Issue #93592, opened 2026-08-24
background_review ignores auxiliary fallback_chain config when primary provider fails


EncoTime:
# Bug Report: background_review ignores auxiliary fallback_chain config

## Summary

When `auxiliary.background_review.provider` fails (e.g. HTTP 429), the per-task `fallback_chain` configured in `auxiliary.background_review.fallback_chain` is never consulted. The review silently fails with `calls=0` instead of falling back to alternative providers.

## Root Cause

`background_review` creates a forked `AIAgent` via `_run_review_in_thread()` and runs `run_conversation()`. The `AIAgent` is instantiated **without** a `fallback_model` parameter, so `agent._fallback_chain` is initialized as `[]`.

When the primary provider fails after retries, `conversation_loop.py` calls `agent._try_activate_fallback()`, which checks `agent._fallback_chain` — finding it empty, it gives up immediately. The auxiliary-specific `_try_configured_fallback_chain()` in `auxiliary_client.py` (which reads `auxiliary.<task>.fallback_chain` from config) is **never called** because `run_conversation` uses the conversation_loop fallback path, not the `call_llm` fallback path.

## Two Independent Fallback Mechanisms

Hermes has two separate fallback systems that don't cross-reference each other:

1. **conversation_loop fallback** (`agent._fallback_chain`)
   - Source: top-level `fallback_providers` in config.yaml
   - Used by: `run_conversation()` → `_try_activate_fallback()`
   - **background_review uses this path**

2. **auxiliary_client.call_llm fallback** (`_try_configured_fallback_chain`)
   - Source: `auxiliary.<task>.fallback_chain` in config.yaml
   - Used by: `call_llm()` → catches errors → tries configured chain
   - **Most other sub-modules use this path** (triage_specifier, goal_judge, kanban_decomposer, etc.)

## Affected Sub-modules

| Sub-module | Call Path | fallback_chain Works? |
|---|---|---|
| triage_specifier | call_llm | ✅ Yes |
| goal_judge | call_llm | ✅ Yes |
| kanban_decomposer | call_llm | ✅ Yes |
| moa_aggregator | call_llm | ✅ Yes |
| curator | call_llm | ✅ Yes |
| title_generation | call_llm | ✅ Yes |
| compression | call_llm | ✅ Yes |
| profile_describer | call_llm | ✅ Yes |
| **background_review** | **run_conversation** | **❌ No** |
| monitor | call_llm | ✅ Yes |
[...]

PR #95101, opened 2026-08-25
feat(authority): manifest schema + compiler + conformance harness

andrexibiza:
## What does this PR do?

This PR adds the first shared **Authority Policy ABI** for Hermes Agent. It gives Python and TypeScript one canonical, fail-closed admission decision before any effect occurs: is this specific operation authorized by this exact policy object and this exact capability proof?

It defines six independent GitHub mutation classes — issue metadata, comments, contents, git data/refs, pull-request creation, and Actions dispatch — so adjacent permissions cannot stand in for the authority required by the actual sink. Each class maps to an explicit broker/direct-symbol sink, required capabilities, allowed actor classes, and allowed resource states.

Both runtimes compile the same canonical manifest and evaluate deterministic refusal precedence. An allowed decision preserves the exact ordered `{capability, source, generation}` proofs used to authorize it and is bound to both `policy_version` and the canonical manifest SHA-256. Callers cannot substitute a widened compiled policy or forged digest while retaining canonical policy identity.

The exact canonical manifest is packaged into both runtimes so later Authority Execution Layer shards consume one immutable policy artifact instead of reparsing a source-tree file or inventing local authority rules. Shared conformance and mutation corpora keep the JSON Schema, Python compiler, TypeScript compiler, and evaluators behaviorally aligned.

This is intentionally **policy admission only**. It does not mint `AuthorityCarrier`s, execute effects, settle operations, or wire mutation sites; those are later shards under #95028.

## Related Issue

Part of #95028. Related architecture/interlocks: #90200, #91911, #93943, #93925.

Independent current-main pressure: #96485 is a live example of the same policy-resolution failure shape at a product sink. Discord `server_actions` currently overloads `None` across absent, malformed, and unavailable configuration states, then downstream consumers interpret that value as unrestricted; schema exposure and runtime enforcement also resolve policy independently. #96485 is **not fixed by this PR** and is not a duplicate of this ABI shard. It is an implementation-side interlock showing why consumers need one typed, fail-closed resolved policy state instead of locally reconstructing authority from lossy sentinels.

## Changed paths

Exactly 10 product/test paths:

- `authority/manifest.v1.json`
- `authority/manifest.schema.json`
- `authority/conformance.v1.json`
- `authority/manifest-mutations.v1.json`
- `hermes_cli/authority/manifest.py`
- `hermes_cli/authority/__init__.py`
- `apps/shared/src/authority/manifest.ts`
- `apps/shared/package.json`
- `tests/authority/test_manifest_conformance.py`
[...]

Issue #95028, opened 2026-08-25
[Architecture] Hermes Authority Execution Layer — the twelve issues are one defect, and the architecture that fixes it


andrexibiza:
# Hermes Authority Execution Layer — the twelve issues are one defect, and the architecture that fixes it

> **Nothing ambient survives a boundary.** A route, profile name, renderer, socket, parent environment, PID, executable path, cached row, current process role, or successful return code may identify a candidate operation. None of those coordinates grants authority. Every consequential transition must consume an explicit current proof object, narrow it to the exact edge, fence it by identity and generation, and produce a typed settlement receipt.

Compressed version of the whole architecture: **every mutation is an edge in a version-controlled authority graph; the edge exists only if the path from admission to it is unbroken; the refusal is the typed reason the path broke.** The AuthorityCarrier is that path. The gate is the path validator. The refusal algebra is the typed output when validation fails.

This is the canonical account of the authority campaign: twelve architecture issues filed by @andrexibiza that are one problem, the six primitives that fix the whole path, the people building it, the PR↔issue interlock (36 PRs), the merge and collision order, the implementation plan (7 PRs), and the adversarial test matrix.

**How to read the numbers.** All issue and PR references are accounted against the 2026-08-25 evidence pull from the `NousResearch/hermes-agent` repository. Origin chains were rebuilt from the captured issue bodies, excluding CI-run references (which share the `#N` shape). PR-to-issue bindings are labeled `closes` only where the PR carries a real `Fixes/Closes` reference, and `related` where the issue number merely appears in the body. That distinction is recorded in the graph, and the honest gaps are listed rather than smoothed over — see the interlock totals in Part VI.

**Evidence boundary.** This is a snapshot-backed public specification, not a claim that GitHub has not changed after the 2026-08-25 evidence pull. The accounting distinguishes the 36 interlock PRs in Part VI from the 89 broader PR references captured in the evidence snapshot.

**Dedup boundary — adjacent surfaces deliberately NOT this issue (do not file here):** #34992 (third-party policy/audit gate proposal for Agent_Sudo), #34320 (deterministic approval-gate feature request), #82591 and #83565 (trackers for Kanban worker authority and child-process credential inheritance — inputs to this architecture, not the same issue). The twelve issues were each separately grounded in real incidents; the PR↔issue interlock audit in Part VI records one core-issue gap (#90149 has no verified binding) and eleven unbound PRs — that is the honest state of the graph, not a claim of completeness.

---

[...]

Issue #90432, opened 2026-08-20
[Feature] Upgrade pre_api_request to a Transform hook — allow plugins to override model/provider/base_url per request


0gl20shk0sbt36:
# [Feature] Upgrade `pre_api_request` to a Transform hook — allow plugins to override model/provider/base_url per request

## Summary

`pre_api_request` currently fires per provider attempt but is **observer-only** — its return value is ignored. I'd like it upgraded to a **Transform** hook (like `pre_transcription` and the `transform_*` family) so plugins can override request-time routing parameters (`model`, `provider`, `base_url`, `api_mode`, and optionally `extra_body`) before the LLM call is dispatched.

### 1. Peak/off-peak provider routing (cost savings)

DeepSeek introduced peak/off-peak pricing (2026-08-17): off-peak is half of peak. For a cache-heavy agent workload (≈99% prompt-cache hit rate, tens of billions of cached input tokens/month), the cheapest channel differs by time of day:

- **Off-peak** → DeepSeek direct API (cache-hit $0.007/M, the lowest cache price anywhere)
- **Peak** → OpenRouter / third-party hosts (flat $0.07–$0.14/M output, no peak multiplier; cheaper than DeepSeek's peak $1.32/M output)

An agent that runs 24/7 can save ~30–40% monthly by routing per-request: DeepSeek direct during off-peak hours, OpenRouter during peak hours. Today this is impossible without either (a) a fragile monkey-patch of `AIAgent._run_agent_loop()`, (b) an external LiteLLM proxy, or (c) a restart-requiring config flip.

### 2. Provider switch aligned with context compaction (zero-cost cache loss)

Switching providers invalidates the prompt cache (a cache lives where it was written). Context compaction also invalidates it. Aligning a provider switch with a compaction event means the cache-miss cost is paid exactly once — the switch becomes effectively free. With a mutating `pre_api_request` hook, a plugin can detect that the previous turn compacted (e.g. via context-length delta or the compaction state exposed to plugins) and switch providers at that exact moment.

### 3. Task-type routing (existing community demand)

The community plugin Hermes ARC / `topic_detect` (routes conversations by topic/domain) currently has to monkey-patch internals because no hook can cleanly change the model before a request (see #23739).

## Proposed change

Make `pre_api_request` accept a return dict that overrides the outgoing request's routing parameters:

```python
def on_pre_api_request(**kwargs):
    # kwargs: task_id, turn_id, api_request_id, session_id, model, provider,
[...]

PR #97168, opened 2026-08-28
feat(plugins): add fail-closed model call policy

tommikkarlsson-a11y:
## Summary

- add a Python-only `pre_model_call_policy` hook at the final provider callback
- use one bounded policy owner for both the primary conversation loop and centralized sync/async auxiliary model calls
- resolve `allow`, `pause`, and `deny` fail closed on timeout, callback error, dispatcher error, or any malformed non-`None` output
- make existing pre-tool policy dispatch exceptions fail closed on both execution paths
- preserve Tommi Karlsson / `tommikkarlsson-a11y` attribution with the normal contributor mapping

## Contract

The policy observes the realized request after request/execution middleware, route selection, and final provider preflight, immediately before provider I/O. A deny or pause makes zero provider calls. Auxiliary denial surfaces directly as `ModelCallPolicyDenied`, bypassing provider retry and fallback.

`None` means no opinion. Every non-`None` result is fully validated before precedence is applied, so malformed allow or lower-priority results cannot bypass fail-closed behavior. The payload carries bounded request identity, exact optional task/mission attribution, sanitized route metadata, and size metadata without prompt content.

The contract covers the host-owned primary conversation path and centralized `call_llm` / `async_call_llm` auxiliary path. Standalone SDK clients outside those paths are explicitly out of scope. Issue #93592 is complementary evidence for the two-path model-I/O architecture.

## Review blockers resolved

- auxiliary sync and async deny witnesses prove zero provider calls
- documented `None` behavior has a direct regression
- malformed allow, malformed lower-priority results, and registration-order permutations all deny
- policy resolution and tool fail-closed dispatch live behind `hermes_cli/model_call_policy.py` instead of regrowing policy semantics in the active godfiles
- branch is restacked onto current upstream `main@dce2ecb8a9428aedf69e959bd15d7a9fa15eae01`
- contributor email mapping is included

## Validation on exact head

- policy/plugin suite: 87 passed
- `tests/run_agent/test_run_agent.py` excluding one environment-dependent Anthropic test: 282 passed
- targeted conversation and tool fail-closed witnesses: 2 passed
- Ruff: passed
- `git diff --check`: passed

The excluded Anthropic test cannot construct its optional client in this local runtime because the `anthropic` package is not installed; the changed policy paths pass on the exact submitted head.

### DIFF
--- a/agent/auxiliary_client.py
+++ b/agent/auxiliary_client.py
@@ -168,6 +168,11 @@ from agent.model_metadata import (
 )
 from hermes_cli.config import get_hermes_home
 from hermes_constants import OPENROUTER_BASE_URL
+from hermes_cli.model_call_policy import (
+    ModelCallPolicyDenied,
+    ModelCallPolicyHalt,
+    enforce_pre_model_call_policy,
+)
 from utils import base_url_host_matches, base_url_hostname, env_float, is_truthy_value, model_forces_max_completion_tokens, normalize_proxy_env_vars
 
 logger = logging.getLogger(__name__)
@@ -3547,6 +3552,44 @@ def _relay_auxiliary_metadata(
     }
 
 
+def _enforce_auxiliary_model_call_policy(
+    client: Any,
+    request: dict[str, Any],
+    *,
+    provider: str | None,
+    api_mode: str | None,
+) -> None:
+    """Authorize one final auxiliary provider attempt."""
+    context = _RELAY_AUX_CALL_CONTEXT.get() or {}
+    attempt_count = max(1, int(context.get("attempt_count") or 0))
+    session_id = ""
+    try:
+        from agent.aux_accounting import get_accounting_context
+
+        accounting = get_accounting_context()
+        if accounting is not None:
+            session_id = str(accounting[1] or "")
+    except Exception:
+        pass
+
+    enforce_pre_model_call_policy(
+        request,
+        call_kind="auxiliary",
+        task_id=os.environ.get("HERMES_KANBAN_TASK", ""),
+        mission_id=os.environ.get("HERMES_KANBAN_MISSION", ""),
+        auxiliary_task=str(context.get("task") or "unknown"),
+        profile=os.environ.get("HERMES_PROFILE", ""),
+        session_id=session_id,
+        api_request_id=str(context.get("request_id") or ""),
+        call_seq=attempt_count,
+        request_attempt=attempt_count - 1,
+        model=str(request.get("model") or context.get("model") or ""),
+        provider=str(provider or context.get("provider") or "auxiliary"),
+        base_url=str(getattr(client, "base_url", "") or ""),
+        api_mode=str(api_mode or context.get("api_mode") or "chat_completions"),
+    )
+
+
 def _relay_sync_completion(
     client: Any,
     kwargs: dict[str, Any],
@@ -3557,17 +3600,27 @@ def _relay_sync_completion(
 ) -> Any:
     callback = create or (lambda request: client.chat.completions.create(**request))
     route = _relay_auxiliary_metadata(provider=provider, api_mode=api_mode)
+
+    def authorized(request: dict[str, Any]) -> Any:
+        _enforce_auxiliary_model_call_policy(
+            client,
+            request,
+            provider=provider,
+            api_mode=api_mode,
+        )
+        return _run_protected_sync_provider_call(callback, request)
+
     # Protected compression calls isolate only the provider callback and stream
     # aggregation.  The owning thread remains free to unwind its lease/DB
     # transaction on hard cancel without touching the process-shared client.
     if route is None:
-        return _run_protected_sync_provider_call(callback, kwargs)
+        return authorized(kwargs)
     provider_name, fallback_model, metadata = route
     from agent import relay_llm
 
     return relay_llm.execute_current(
         kwargs,
-        lambda request: _run_protected_sync_provider_call(callback, request),
+        authorized,
         name=provider_name,
         model_name=str(kwargs.get("model") or fallback_model),
         metadata=metadata,
@@ -3585,14 +3638,24 @@ async def _relay_async_completion(
 ) -> Any:
     callback = create or (lambda request: client.chat.completions.create(**request))
     route = _relay_auxiliary_metadata(provider=provider, api_mode=api_mode)
+
+    async def authorized(request: dict[str, Any]) -> Any:
+        _enforce_auxiliary_model_call_policy(
+            client,
+            request,
+            provider=provider,
+            api_mode=api_mode,
+        )
+        return await callback(request)
+
     if route is None:
-        return await callback(kwargs)
+        return await authorized(kwargs)
     provider_name, fallback_model, metadata = route
     from agent import relay_llm
 
     return await relay_llm.execute_current_async(
         kwargs,
-        callback,
+        authorized,
         name=provider_name,
         model_name=str(kwargs.get("model") or fallback_model),
         metadata=metadata,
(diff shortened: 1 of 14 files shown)


andrexibiza (commented):
Reviewed exact head `1704b57064db2bfba88e7d07d4bb037b45d9e24d` (tree `c29de8e7da3afe5d4a86ce3d6dd90705ea9c50db`). The core direction is good: putting a synchronous gate after request shaping and before observability/provider I/O is the right place for policy, and the paired `pre_tool_call` exception change correctly removes an existing fail-open path. The timeout/exception cases are also tested at the resolver boundary. There are three correctness blockers plus landing/ownership gates before this can safely carry the documented contract.

### P1 — “before every provider attempt” currently has a large ungated auxiliary path

The docs added by this PR say `pre_model_call_policy` fires before **every provider attempt**, and the PR motivation explicitly calls out model budgets / durable run state. The implementation is only in `agent/conversation_loop.py::run_conversation()`.

Hermes has a second independent model-I/O path in `agent/auxiliary_client.py`: `call_llm()` and its relay helpers ultimately call `client.chat.completions.create(**request)` (plus Codex/Responses adapters) without passing through `conversation_loop.py`, and this exact head contains no `pre_model_call_policy` acquisition in that file. Live issue #93592 independently documents the same two-path architecture: background review uses `run_conversation`, while triage, goal judge, Kanban decomposition, MoA aggregation, curator, title generation, compression, profile description, monitor, and MoA reference use `auxiliary_client.call_llm`.

That means a budget/durable-run plugin can deny the main conversation and Hermes can still spend model authority through auxiliary calls in the same process. This is the other side of the shape, not a documentation nit.

Required landing shape: put the policy acquisition at a shared provider-attempt authority boundary used by both conversation and auxiliary dispatch (or explicitly narrow the contract and name so it cannot be mistaken for a process-wide model-call gate). Add at least one adversarial auxiliary witness proving a deny decision results in **zero provider calls**. #93592 is complementary evidence and should be linked rather than absorbed or credited away.

### P1 — the documented `None` result is treated as a denial

`hermes_cli/plugins.py::resolve_pre_model_call_policy()` documents:

> `None` means the callback is not applicable.

But the resolver currently starts each result with:

```python
if not isinstance(result, dict):
    return _fail_closed_hook_result("pre_model_call_policy")
```

So a perfectly valid callback that follows the new documented contract and simply `return None` produces a fail-closed denial. `invoke_hook()` preserves callback returns in the result list, so this is reachable as written.

Required fix: consume `None` as “no opinion” *before* malformed-result handling, and add the direct regression. This matters especially with multiple policy plugins, where one non-applicable plugin must not veto another valid decision.

### P1 — malformed `allow` results can pass because validation happens after precedence

The resolver promises that any malformed non-`None` result denies. However message validation occurs only after:

```python
if rank[action] <= rank[decision["action"]]:
    continue
```

The initial decision is `allow`, so a callback returning e.g. `{"action": "allow", "message": object()}` is skipped before the malformed `message` is checked and the provider request is allowed. The same ordering can hide malformed lower/equal-priority results once a stronger decision has already been seen.

Required fix: validate the **entire result shape** (`action`, `message`, and any future schema fields) before applying precedence. Add regressions for malformed `allow`, malformed lower-priority results after `pause`/`deny`, and registration-order permutations. Fail-closed precedence should choose among valid decisions; it must never double as a validation bypass.

### Hard landing gates — ownership, attribution, and exact-head evidence

This PR adds behavior directly back into three active no-regrowth/fracture surfaces: `agent/conversation_loop.py` (#78641; the changed hunk is already around line 6.7K), `agent/tool_executor.py` (#79975), and `hermes_cli/plugins.py` (#79969; resolver hunk around line 6.5K). #78647 records the repository law that completed godfile decomposition is monotonic: future behavior belongs behind bounded owners, not by regrowing the historical file. Please preserve this feature, but restack it through the owning shard seams / explicit merge order instead of making three policy owners grow in place.

There is also a direct adjacent boundary with #90432: if `pre_api_request` becomes a request transform, the semantic order must remain **final route/request transform → policy proof over that exact realized destination → provider I/O**. A policy decision made before a later model/provider/base-URL transform would authorize the wrong sink. #95028 / #95101 are broader authority-policy work and are adjacent rather than duplicate; this hook does not need to become that whole architecture, but it should not create an incompatible second meaning for fail-closed policy identity/refusal.

Attribution is not landing-ready yet: the sole commit is authored as `Tommi Karlsson <tommi.k.karlsson@gmail.com>`, while current `main` has no `contributors/emails/tommi.k.karlsson@gmail.com` mapping. Preserve Tommi Karlsson / `tommikkarlsson-a11y` credit by adding the normal mapping rather than rewriting history.

Finally, exact-head hosted evidence is not green: CI `33173979616`, Docker `33173979033`, and Nix `33173978960` are all `action_required`, and each currently has zero jobs. The PR is also now 1 commit ahead / 12 commits behind live `main@eff97a8a0545615010cb050e7cd76ae258bd0d23` from merge base `9978706e9303dbf990d90e744b131361449d73b9`. The local 79 + 8 + partial run-agent results are useful development evidence, but not an exact-head acceptance receipt.

The strong parts are worth keeping: the policy is intentionally separate from the fail-open context hook, the provider call is fenced before observability/debug I/O on the conversation path, the `pre_tool_call` dispatcher exception is now fail-closed in both execution paths, and timeout handling is explicit rather than implicit. Close the two resolver holes, make the model gate cover the actual model-I/O graph, land it behind bounded owners with clean attribution, then rerun every submitted commit on current main. That gives this a much stronger claim than “new hook”: it becomes a real fail-closed model-call authority boundary. 🚀

tommikkarlsson-a11y:
Restacked onto current `main@dce2ecb8a9428aedf69e959bd15d7a9fa15eae01`; exact head is `e33c124e9e76cfb7c1fc57c2e4d7eb33d1a27c76`.

The three P1 correctness blockers are addressed: the shared gate now covers centralized sync/async auxiliary calls with zero-provider-call denial witnesses, `None` is treated as no opinion, and every non-`None` result is fully validated before precedence. Policy logic is behind the bounded `hermes_cli/model_call_policy.py` owner, the contributor mapping is present, and exact-head local validation is green (87 policy/plugin tests, 282 run-agent regressions, Ruff, and diff check).

The hosted CI, Docker, and Nix workflows still end as `action_required` with zero jobs. Maintainer workflow approval and re-review are the remaining landing gates.

agent/auxiliary_client.py
```diff
@@ -3557,17 +3600,27 @@ def _relay_sync_completion(
...
     route = _relay_auxiliary_metadata(provider=provider, api_mode=api_mode)
+
+    def authorized(request: dict[str, Any]) -> Any:
+        _enforce_auxiliary_model_call_policy(
+            client,
+            request,
+            provider=provider,
+            api_mode=api_mode,
+        )
+        return _run_protected_sync_provider_call(callback, request)
```
andrexibiza, line 3611: **P1 — one policy acquisition can still authorize two provider attempts.**

For progress-hooked synchronous calls, `callback` is `_create_with_progress`. This line acquires policy once for the original `request`, then hands control to that helper. `_create_with_progress` can send `client.chat.completions.create(**stream_kwargs)`, catch a non-transient provider rejection of streaming, and immediately send `client.chat.completions.create(**kwargs)` as a non-stream fallback. That second provider I/O never returns through `authorized`, so `pre_model_call_policy` is not invoked again, `request_attempt` does not advance, and an allow for attempt 0 silently authorizes attempt 1 even when policy would deny it. The async force-stream helper has the same sequencing problem: it constructs the realized `stream_kwargs` after this gate.

Move acquisition into the actual dispatch primitive used for each `client.chat.completions.create(...)`, or make the stream helpers call an authorized dispatch separately for every realized stream/plain request. Add an adversarial sync witness where policy returns `allow` then `deny`, the streamed provider request is rejected, and the non-stream fallback makes no second provider call; assert policy-call count and `request_attempt` remain one-to-one with provider attempts.

andrexibiza (commented):
Re-reviewed exact head `e33c124e9e76cfb7c1fc57c2e4d7eb33d1a27c76` against current `main@dce2ecb8a9428aedf69e959bd15d7a9fa15eae01`.

The original three P1s are resolved on this object: centralized auxiliary direct-deny witnesses now prove zero provider calls; `None` is consumed as no opinion; and every non-`None` result is fully validated before precedence. Policy resolution and the adjacent tool fail-closed semantics now live behind `hermes_cli/model_call_policy.py`, and the contributor mapping is present.

One P1 remains at the concrete auxiliary provider-attempt boundary; see the inline finding. Until every actual provider dispatch reacquires policy, the advertised fail-closed per-attempt contract is not true.

Exact-head hosted acceptance is also still absent: [CI 33316390291](https://github.com/NousResearch/hermes-agent/actions/runs/33316390291), [Docker 33316389929](https://github.com/NousResearch/hermes-agent/actions/runs/33316389929), and [Nix 33316389986](https://github.com/NousResearch/hermes-agent/actions/runs/33316389986) are each `action_required` with zero jobs. The reported local suites are useful development evidence, but they are not an exact-head hosted acceptance receipt.

PR #75109, opened 2026-07-31
fix(cli): prevent python -m main double-import profile rehoming

weddle:
## What does this PR do?

`python -m hermes_cli.main` registers the running module only as `__main__`. When code in the same process later performs a canonical `import hermes_cli.main`, Python finds no cached entry under that name and executes the whole file a second time.

Several handlers import it that way at runtime. `setup.status` and `setup.runtime_check` do (`tui_gateway/methods_config.py:354,375`). So does the argv builder for `/api/pty` (`hermes_cli/web_server.py:14545`), and plugin aux-task registration (`hermes_cli/plugins.py:1130`).

The second execution corrupts process state. On the first pass, `_apply_profile_override()` consumes `-p <name>` from `sys.argv` (`main.py:678-681`). The rerun finds no flag, falls back to the sticky `active_profile`, and rewrites `os.environ["HERMES_HOME"]` (`main.py:656-677`). A serve launched with `-p default` moves to the sticky profile's home the moment a desktop client asks for runtime readiness. Globals captured at import time keep the original value (`tui_gateway/server.py:54`), so one process resolves profiles two different ways. The rerun also calls `load_hermes_dotenv(override=True)` again, overwriting env vars set after startup.

I observed this on a live serve started with `-p default` while the sticky `active_profile` named a second profile (`alpha` here). `projects.list` returned the alpha store for `profile:"default"` and for unknown profile names. Omitting the param gave the same result. `/api/profiles/active` reported `current: alpha`.

The fix registers the running module under its canonical name so later imports reuse it, and binds it as `main` on the loaded parent package (a `sys.modules` cache hit skips importlib's parent-attribute step, so the alias alone would leave the dotted attribute `hermes_cli.main` unset):

```python
if __name__ == "__main__" and __spec__ is not None:
    sys.modules.setdefault(__spec__.name, sys.modules[__name__])
    _canonical = sys.modules[__spec__.name]
    _parent_name, _, _child = __spec__.name.rpartition(".")
    _parent = sys.modules.get(_parent_name) if _parent_name else None
    if _parent is not None and not hasattr(_parent, _child):
        setattr(_parent, _child, _canonical)
```

With the rerun gone, every top-level side effect is covered, including the dotenv reload.

Open PRs for #66907:

- #67048 wraps `_apply_profile_override()` in a PID-scoped sentinel. The HERMES_HOME rewrite stops. The rest of the module top level still reruns on canonical import.
[...]

PR #67048, opened 2026-07-18
fix: add PID-scoped run-once sentinel to _apply_profile_override (#66907)

webtecnica:
## Summary

Under `python -m hermes_cli.main --profile default gateway run`, the module executes twice: once as `__main__`, and again when any code later does `import hermes_cli.main`. The first pass parses `--profile default`, sets `HERMES_HOME`, and **strips the flag from `sys.argv`**. The second pass then sees no `--profile` flag and falls through to the sticky `active_profile` file — silently rebinding `HERMES_HOME` to a different profile mid-startup.

## Fix

Add a PID-scoped run-once sentinel at the top of `_apply_profile_override()`:

```python
_sentinel_key = "_HERMES_PROFILE_OVERRIDE_APPLIED"
if os.environ.get(_sentinel_key) == str(os.getpid()):
    return
os.environ[_sentinel_key] = str(os.getpid())
```

PID-scoping keeps spawned children resolving their own profile fresh (they inherit the env var but have a different PID).

## Changes

- **`hermes_cli/main.py`**: Add PID-scoped sentinel guard at the top of `_apply_profile_override()`
- **`tests/hermes_cli/test_apply_profile_override.py`**: Clear the sentinel before each call to `_apply_profile_override()` so tests aren't affected by the PID guard in same-process pytest runs

## Verification

All 12 existing tests in `test_apply_profile_override.py` pass with the sentinel in place.

Fixes #66907

PR #97174, opened 2026-08-28
fix(cli): register canonical sys.modules alias under runpy to stop profile re-scope

criptoboi69:
## What does this PR do?

Fixes a silent profile re-scope bug when Hermes is launched via `python -m hermes_cli.main` (the systemd unit's `ExecStart`).

`runpy` executes the module as `__main__` **without** registering it under its canonical name in `sys.modules`. The first lazy `from hermes_cli.main import ...` in the process — `gateway/code_skew.py`'s boot fingerprint, the first such import in a gateway process — therefore re-executed the entire module, running `_apply_profile_override()` a second time against the already-stripped `argv`. With a sticky `active_profile` present, that re-resolved `HERMES_HOME` to the named profile and silently re-scoped a `-p default` gateway to that profile, dropping the multiplexer's port-binding platforms (`api_server` 8642) after every in-process restart.

The fix registers the canonical alias in the `__main__` guard so lazy imports resolve to the already-executed module instead of re-running it.

## Related Issue

N/A (no existing issue found for this symptom).

## Type of Change

## Changes Made

- `hermes_cli/main.py`: in the `if __name__ == "__main__":` guard, add `sys.modules.setdefault("hermes_cli.main", sys.modules["__main__"])` before `main()` (15 lines with explanatory comment).
- `tests/hermes_cli/test_apply_profile_override.py`: regression test `TestRunpyModuleExecutionRegistersCanonicalAlias` (72 lines) that reproduces runpy execution + lazy import.

## How to Test

1. Reproduce (before fix): run the module via runpy with a sticky `active_profile` set and a `-p default` argv, then trigger a lazy `from hermes_cli.main import ...` — `HERMES_HOME` gets re-resolved to the named profile (e.g. `<root>/profiles/<sticky>`).
2. With the fix, the lazy import resolves to the already-executed module and `HERMES_HOME` stays `<root>`.
3. Regression test: `pytest tests/hermes_cli/test_apply_profile_override.py -k RunpyModuleExecution` — fails without the fix (RED verified by reverting the guard), passes with it (GREEN).

Empirical before/after on a real gateway process: before the fix `Session storage: <root>/profiles/<sticky>/sessions` + `Active profile: <sticky>`; after the fix `Session storage: <root>/sessions` and no `Active profile` line.

### Code

- [x] I've tested on my platform: Fedora 44 (Linux), Python 3.13

### DIFF
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -14782,4 +14782,19 @@ def main():
 
 
 if __name__ == "__main__":
+    # ``python -m hermes_cli.main`` (runpy) executes this module as
+    # ``__main__`` WITHOUT registering it under its canonical name in
+    # sys.modules. The first lazy ``from hermes_cli.main import ...`` in the
+    # process (gateway/code_skew.py's boot fingerprint is the first such
+    # import in a gateway process) would then re-execute the entire module —
+    # running _apply_profile_override() a second time against the
+    # already-stripped argv. With a sticky active_profile present, that
+    # re-resolves HERMES_HOME to the named profile and silently re-scopes a
+    # ``-p default`` gateway (the systemd unit's ExecStart) to that profile,
+    # losing the multiplexer's port-binding platforms (api_server 8642) after
+    # every in-process restart. Register the canonical alias so lazy imports
+    # resolve to this already-executed module instead of re-running it.
+    import sys as _sys
+
+    _sys.modules.setdefault("hermes_cli.main", _sys.modules["__main__"])
     main()
--- a/tests/hermes_cli/test_apply_profile_override.py
+++ b/tests/hermes_cli/test_apply_profile_override.py
@@ -12,7 +12,9 @@ active_profile (child-process inheritance contract).
 from __future__ import annotations
 
 import os
+import subprocess
 import sys
+import textwrap
 from pathlib import Path
 from types import SimpleNamespace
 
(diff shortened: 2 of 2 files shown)


alt-glitch:

Duplicate of #75109: both patches alias the runpy __main__ module under hermes_cli.main to prevent the second import and profile re-scope. #75109 is the earlier, more complete open implementation.

criptoboi69:
Thanks for the triage flag. Acknowledged — this is the same structural fix as #75109 (canonical `sys.modules` alias under the `__main__` guard), and I was not aware of that PR when opening this one.

What this PR adds on top of the existing cluster:

1. **Production validation.** This exact fix is deployed and validated in production (systemd gateway, `python -m hermes_cli.main` ExecStart): 13/13 verification checks passed, with empirical before/after on a live gateway process — before: `Session storage: <root>/profiles/<sticky>/sessions` + `Active profile: <sticky>`; after: `Session storage: <root>/sessions`, no `Active profile` line.
2. **Minimal variant.** 15 lines in `hermes_cli/main.py` (vs 29 in #75109) — the `setdefault` one-liner plus an explanatory comment, no parent-attribute rebinding.
3. **Runpy-specific regression test.** `TestRunpyModuleExecutionRegistersCanonicalAlias` executes the module through the real runpy path (`exec` as `__main__` + lazy `from hermes_cli.main import ...`) and asserts the canonical alias resolves to the already-executed module — RED/GREEN verified on current `main`.

I'm happy to close this PR in favor of #75109 if maintainers prefer the earlier implementation, and to contribute the runpy regression test there. Keeping it open for now so the production evidence is visible in the decision between #67048 and #75109.

Enough1122:

Accurate diagnosis and a correct, minimal fix. The runpy behavior is real — `python -m hermes_cli.main` executes the module as `__main__` without registering the canonical name, so the first lazy `from hermes_cli.main import ...` re-runs the whole module and `_apply_profile_override()` a second time, silently re-scoping a `-p default` gateway. The alias registration in the `__main__` block is the right seam: it runs before `main()`, `setdefault` won't clobber a canonical module that legitimately pre-exists, and it also covers direct script execution (`python hermes_cli/main.py`), which has the same no-canonical-alias behavior. The failure-chain comment is excellent.

The one thing I'd strengthen is the test:

1. **`rsplit("    main()", 1)[0] + "    pass\n"` couples the test to source layout.** It neutralizes the trailing `main()` by exact string/indentation match. If the `__main__` block ever changes indentation, or a `main()`-calling line with different indentation appears later in the file, the neutralization either misses or hits the wrong call. A more robust approach that avoids touching source text at all: inject `main = lambda: None` into the exec globals (the test already updates `__main__.__dict__`), so the trailing `main()` call becomes a harmless no-op. That keeps the "execute real module source as `__main__`, then lazy-import" semantics intact without the fragile source-shape dependency — which also sidesteps the repo's "don't depend on source shape in tests" anti-pattern.

2. **The fix is `hermes_cli/main.py`-specific; other `__main__`-executed entry modules share the hazard.** Any module run via `python -m <pkg>` and later imported by canonical name has the same double-execution profile. The lazy import that trips it (code_skew boot fingerprint) is specific to the gateway, so this PR is correctly scoped to the reported bug — but a shared one-line helper (e.g. `register_canonical_alias(__name__)` used by any `__main__` entry) or a note in the comment would make the pattern discoverable for the next entry module that hits it.

3. Minor: the test imports the private `_read_git_revision_fingerprint` (verified present at hermes_cli/main.py:841) to simulate the first lazy import. That's faithful to the real code_skew trigger, but it means a rename of that symbol breaks the test without signaling why. Acceptable as-is; just flagging the coupling.

The subprocess E2E shape (real source executed as `__main__`, real lazy import, asserting `HERMES_HOME` isn't re-scoped) is exactly the right kind of test for this bug — real resolution chain, no mocks. Good work; only the `rsplit` fragility is worth adjusting.

PR #97175, opened 2026-08-28
fix(desktop): preserve two-space hard breaks in preprocessMarkdown

### LINKED ISSUE #97117: [Bug]: Desktop Markdown preprocessor strips two-space hard breaks
### Bug Description

Hermes Desktop removes standard Markdown two-space hard breaks before the text reaches Streamdown.

A real assistant response contained separate source lines ending in two literal spaces followed by `\n`. The stored message was correct, but Desktop displayed the entries as one paragraph. The TUI and classic CLI did not show the same collapse.

This is deterministic rather than intermittent: `preprocessMarkdown()` removes the two spaces that carry Markdown semantics.

### Steps to Reproduce

1. Pass the following string through the Desktop Markdown rendering path:

   ```ts
   const input = "First line  \nSecond line"
   ```

   The two spaces before `\n` are literal and constitute a Markdown hard break.

2. Inspect the preprocessed output:

   ```ts
   preprocessMarkdown(input)
   // Current output: "First line\nSecond line"
   ```

3. Render the result in a Hermes Desktop assistant message.

A real-world reproduction is a source list formatted this way:

```text
[1] First source  
[2] Second source  
[3] Third source
```

The message stored in the session database retains the two trailing spaces and physical newlines, but Desktop displays the entries as a single paragraph. This confirms that the content is not being collapsed during model generation or persistence.

### Expected Behavior

Desktop should preserve standard Markdown hard-break semantics. The following input:

```text
"First line  \nSecond line"
```

should render as two visible lines, equivalent to:

```html
First line<br>
Second line
```

Ordinary CommonMark soft breaks should retain their normal semantics; the fix should not require treating every single newline as a hard break.

### Actual Behavior

`preprocessMarkdown()` changes:

```text
"First line  \nSecond line"
```

into:

```text
"First line\nSecond line"
```

Streamdown then interprets the remaining single newline as a CommonMark soft break, so the lines appear in one paragraph:

```text
First line Second line
```

### Affected Component

Other — Hermes Desktop Markdown rendering (`apps/desktop`)

The TUI and classic CLI use different rendering paths and serve as working controls:

- TUI splits the source into physical lines in `ui-tui/src/components/markdown.tsx`.
- Classic CLI renders through Rich Markdown and does not use the Desktop preprocessor.

N/A — local Hermes Desktop chat

### Debug Report

```shell
Not uploaded: this is a deterministic renderer-only reproduction that does not depend on provider, gateway, configuration, or logs.
```

### Operating System

macOS 26.6.2 (25G83)

### Python Version

Hermes runtime: Python 3.11.15

### Hermes Version

Hermes Agent v0.20.6 (2026.8.27); reproduced on checkout `48d25280669f645550c86b7540c01996f611be63`

```shell
No exception is raised. The failure is incorrect Markdown preprocessing and rendering.
```

The final step of `preprocessMarkdown()` in `apps/desktop/src/lib/markdown-preprocess.ts` contains:

```ts
.join('')
.replace(/[ \t]+\n/g, '\n')
```

The regular expression removes all trailing spaces and tabs before a newline. In Markdown, two or more spaces immediately before a newline are not insignificant whitespace; they encode a hard line break. The cleanup therefore changes the document's meaning before Streamdown parses it.

`git blame` traces this line to the initial Desktop merge, PR #20059 (`Add Hermes desktop app`), commit `51c68d4ab1a9e3c62fb1048fccb84144c409f0e7`.

No focused regression test for two-space hard breaks was found in the Desktop Markdown test suites.

### Proposed Fix (optional)

Make trailing-whitespace normalization Markdown-aware:

- Do not remove a run of two or more spaces before `\n` in Markdown prose, or normalize such a run to exactly two spaces.
- Preserve fenced-code content unchanged.
- Keep ordinary soft-break behavior unchanged.
- Add a regression test asserting that `preprocessMarkdown("First line  \nSecond line")` retains hard-break semantics.

A CSS-only `white-space: pre-wrap` workaround would make all soft breaks visible and would change normal CommonMark paragraph behavior, so preserving the semantic marker in preprocessing is the narrower fix.

### Related Issues and PRs

- #95995 reports intermittent multiline collapse and code-block styling loss, but this issue is a separate deterministic hard-break preprocessing bug with a specific root cause.
- #70561 adds `whitespace-pre-wrap` to prose and may mask the symptom, but it does not preserve Markdown hard-break semantics and currently has unrelated build/test failures.
- #49574 preserves line formatting inside prose-like plaintext fences only; normal Markdown paragraphs remain affected.

### Acceptance Criteria

- [ ] Two-space Markdown hard breaks survive Desktop preprocessing and render as visible line breaks.
- [ ] Ordinary single-newline soft breaks retain CommonMark behavior.
- [ ] Fenced-code contents are not modified.
- [ ] A focused `markdown-preprocess` regression test covers the two-space hard-break case.

Not committing to a PR at this time; the deterministic reproduction and source-level root cause are provided to make the issue directly actionable.

Jackal991:
## Summary

Hermes Desktop destroyed CommonMark two-space hard breaks before the text reached Streamdown: the trailing-whitespace cleanup ran `/[ \t]+\n/ → '\n'` in two places inside the rendering pipeline, so `"First line  \nSecond line"` rendered as one paragraph (`First line Second line`) while TUI and classic CLI rendered it correctly.

Both sites are now Markdown-aware: a run of **two or more spaces** directly before a newline survives byte-identical (that run *is* the hard break), while a lone trailing space, any tab, or junk ahead of a hard-break run still collapses exactly as before. A newline with no trailing whitespace never matches, so ordinary soft breaks keep their normal semantics — no change to single-newline behavior.

Empirically confirmed against Streamdown's own remark stack (remark-parse 11 / micromark): raw two-space input parses to `breakCount: 1`; the old preprocessor's output parses to `breakCount: 0` (fused into one text run); the fixed output restores `breakCount: 1`.

## Changes

- `apps/desktop/src/lib/markdown-preprocess.ts` — the final `.replace(/[ \t]+\n/g, '\n')` of `preprocessMarkdown()` is replaced with a Markdown-aware pass: `/[ \t]*(  +)\n|[ \t]+\n/g` keeps a 2+ space run before `\n` byte-identical and still collapses everything else at end-of-line exactly as before.
- `apps/desktop/src/lib/preview-targets.ts` — `stripPreviewTargets()` had the *same* strip and runs **inside** `preprocessMarkdown()` on every prose segment; without the same fix there, hard breaks still vanished (found by the regression tests failing after fixing only the final pass). It has exactly one production caller: `preprocessMarkdown`.
- `apps/desktop/src/components/assistant-ui/markdown-text.test.ts` — three focused regression tests: two-space hard break passes through byte-identical; hard breaks survive between source-list entries (`[1] …  \n[2] …`); a single trailing space is still stripped (original cleanup behavior kept).

## Tests

- Regression tests **fail against the old code** (2 failed | 48 passed — verified by temporarily restoring the old regexes at both sites) and **pass with the fix** (50/50).
- Full desktop ui suite: **618 files / 6131 tests pass**; `npm run typecheck` clean across all three tsconfig projects; `preview-targets.test.ts` passes unchanged.
- Focused: `vitest run --project ui src/components/assistant-ui/markdown-text.test.ts src/lib/preview-targets.test.ts` — all pass.

Closes #97117

## Notes

Three *conditional* message-ingestion normalizers (`renderMediaTags`, `normalizeCleanedText` in embedded-images, `stripGeneratedImageEchoes`) also collapse trailing whitespace, but they sanitize stored content on specific content triggers (MEDIA tags, data-image payloads, generated-image echoes) with different normalization semantics — deliberately left untouched here; happy to follow up if maintainers want the same treatment there.

### DIFF
--- a/apps/desktop/src/components/assistant-ui/markdown-text.test.ts
+++ b/apps/desktop/src/components/assistant-ui/markdown-text.test.ts
@@ -102,6 +102,29 @@ describe('preprocessMarkdown', () => {
     expect(output).toContain('`items[0]`')
   })
 
+  it('preserves two-space hard breaks through preprocessing', () => {
+    // Two trailing spaces before a newline are a CommonMark hard break —
+    // Streamdown renders them as <br>. Preprocessing must be byte-identical
+    // here or the break silently disappears and both lines fuse into one.
+    const input = 'First line  \nSecond line'
+
+    expect(preprocessMarkdown(input)).toBe(input)
+  })
+
+  it('keeps hard breaks between source-list entries', () => {
+    const input = ['[1] First source  ', '[2] Second source  ', '[3] Third source'].join('\n')
+
+    const output = preprocessMarkdown(input)
+
+    expect(output).toContain('First source  \n[2]')
+  })
+
+  it('still strips a single trailing space before a newline', () => {
+    // Incidental trailing whitespace is not a hard break — the cleanup
+    // this pass was written for must keep working.
+    expect(preprocessMarkdown('First line \nSecond line')).toBe('First line\nSecond line')
+  })
+
   it('demotes title/url blocks wrapped in malformed inline fences', () => {
     const input = [
       '**🚢 TOMORROW (Fajardo, crystal clear cays, pickup avail):**',
--- a/apps/desktop/src/lib/markdown-preprocess.ts
+++ b/apps/desktop/src/lib/markdown-preprocess.ts
@@ -664,7 +664,17 @@ export function preprocessMarkdown(text: string): string {
       return leading + transformed + trailing
     })
     .join('')
-    .replace(/[ \t]+\n/g, '\n')
+    // Trailing-whitespace cleanup, Markdown-aware: two or more spaces
+    // directly before a newline are a CommonMark hard break — Streamdown
+    // renders them as <br>, so that run must reach it byte-identical or
+    // two lines fuse into one paragraph. Everything else at end-of-line
+    // (a lone space, any tab, or junk ahead of a hard-break run) is
+    // incidental trailing whitespace and still collapses exactly as this
+    // pass always did. A newline with no trailing whitespace never matches,
+    // so single newlines keep their soft-break meaning.
+    .replace(/[ \t]*(  +)\n|[ \t]+\n/g, (_match: string, hardBreak: string | undefined) =>
+      hardBreak ? `${hardBreak}\n` : '\n'
+    )
 }
 
 /**
--- a/apps/desktop/src/lib/preview-targets.ts
+++ b/apps/desktop/src/lib/preview-targets.ts
@@ -3,7 +3,14 @@ const PREVIEW_MARKDOWN_RE = /\[Preview:[^\]]+\]\((?<href>#preview[:/][^)]+)\)/gi
 export function stripPreviewTargets(text: string): string {
   return text
     .replace(PREVIEW_MARKDOWN_RE, '')
-    .replace(/[ \t]+\n/g, '\n')
+    // Trailing-whitespace cleanup, Markdown-aware: two or more spaces
+    // directly before a newline are a CommonMark hard break and must reach
+    // Streamdown byte-identical or two lines fuse into one paragraph. A
+    // lone space, any tab, or junk ahead of a hard-break run is incidental
+    // trailing whitespace and still collapses as this pass always did.
+    .replace(/[ \t]*(  +)\n|[ \t]+\n/g, (_match: string, hardBreak: string | undefined) =>
+      hardBreak ? `${hardBreak}\n` : '\n'
+    )
     .replace(/\n{3,}/g, '\n\n')
     .trim()
 }


Enough1122:

**Overall:** Fixes CommonMark hard-break preservation through markdown preprocessing (#97214 sibling).

**What it does**
- `apps/desktop/src/lib/markdown-preprocess.ts` changes `/[ \t]+\n/g` → markdown-aware replacement that preserves `  \n` / `   \n` (two+ spaces before newline → `<br>`) and only strips other trailing whitespace; same fix applied to `stripPreviewTargets` in `preview-targets.ts`.
- `apps/desktop/src/components/assistant-ui/markdown-text.test.ts` adds `preserves two-space hard breaks through preprocessing` case ensuring `First line  \nSecond line` survives byte-identical to Streamdown.

**Non-blocking notes**
- Regex now distinguishes "two+ spaces" general - a single trailing space before newline is still stripped — correct per spec (single space is not a hard break), but if user intentionally uses single-space + newline for soft break, behavior is unchanged (was stripped before too).
- `stripPreviewTargets` applies two regexes in sequence; order matters — verify hard-break preservation runs after preview strip, not before, so stripped preview match doesn't leave stray spaces that get preserved as false hard break.

Correct, minimal fix for fused-line bug.

Jackal991:
Agreed on the overall — the regex now distinguishes the two+ space hard-break run from incidental trailing whitespace, and the single-space case is correctly still stripped (a lone space isn't a hard break per CommonMark).

On the ordering note: it's already satisfied. In `stripPreviewTargets` the preview-strip regex runs first, then the whitespace cleanup — so a stripped preview match can't leave stray spaces that get preserved as a false hard break. The two regexes are applied in sequence in the source, preview strip before the trailing-whitespace pass, and the `keeps hard breaks between source-list entries` test exercises the real path. Nothing to change.

teknium1:
Thanks @Jackal991 for this contribution. This fix path is now on main via #104980 (https://github.com/NousResearch/hermes-agent/pull/104980), merge commit `583ff8b6b149b45e65de375977e92684ae022611`.

Hard-break direction superseded by deleting destructive whitespace cleanup across both original and sibling sites; single-space-strip assertion intentionally replaced by semantic behavior.

Closing this alternative as superseded by the verified merged implementation, not rejecting the reported problem. See the replacement PR for exact live before/after evidence and remaining fidelity limits. Other unresolved issue variants remain open.

PR #97176, opened 2026-08-28
fix(profiles): tolerate unreadable skill trees

### LINKED ISSUE #97173: profiles.list fails when skill scan hits an unreadable Windows junction
## Bug description

On Windows, profile enumeration can fail entirely when presentation-only skill counting encounters an unreadable junction, reparse point, or other inaccessible subtree under a profile's `skills` directory.

The exception propagates from `hermes_cli.profiles._count_skills()` through `list_profiles()`. That can break `hermes profile list`, the `profiles.list` gateway method, and the Desktop Bots roster even though the profiles and backend are otherwise healthy.

A real failure produced:

```text
OSError: [WinError 448] The path cannot be traversed because it contains an untrusted mount point
```

## Steps to reproduce

1. On Windows, create or expose a profile whose `%HERMES_HOME%\profiles\<profile>\skills` tree contains one readable skill and one junction/reparse subtree that the current process cannot traverse.
2. Call `hermes_cli.profiles.list_profiles()` (or invoke the gateway `profiles.list` path).
3. Observe that `Path.rglob("SKILL.md")` raises `OSError` and the whole profile inventory fails.

A deterministic regression test can model the Windows condition by making `os.scandir()` raise `OSError(448, ...)` for one child under the skills tree. Current `main` fails that test while traversing `Path.rglob()`.

## Expected behavior

- The profile remains in the inventory.
- Readable skills are still counted.
- An unreadable subtree degrades presentation metadata instead of hiding every profile.

## Actual behavior

The `OSError` escapes `_count_skills()` and aborts `list_profiles()`, making the roster unavailable.

## Root cause

Current `main` uses a direct recursive scan in `hermes_cli/profiles.py`:

```python
for md in skills_dir.rglob("SKILL.md"):
    ...
```

It also calls `skills_dir.is_dir()` without guarding `OSError`.

Hermes already has the canonical `agent.skill_utils.iter_skill_index_files()` walker. It applies the shared exclusion/org/support-directory rules and uses `os.walk()`, which skips inaccessible subtrees by default. Profile skill counting should use that helper and fail closed to a count of zero on an unexpected `OSError`, rather than failing the profile roster.

## Environment

- Windows 11 (`Microsoft Windows NT 10.0.26200.0`)
- Python 3.11.15
- Reproduced in Hermes Agent v0.20.6 (2026.8.27)
- Confirmed the vulnerable `Path.rglob()` implementation is still present on `main` at `9978706e9303dbf990d90e744b131361449d73b9`

## Proposed fix

- Reuse `iter_skill_index_files()` in `_count_skills()`.
- Guard root existence and scan operations against `OSError`.
- Add a regression test proving that one blocked subtree does not hide the profile and that readable skills are still counted.

toughCSB:
## What does this PR do?

Keeps profile enumeration available when skill-counting encounters an unreadable Windows junction, reparse point, or other inaccessible subtree.

`skill_count` is presentation metadata, so an inaccessible subtree should not make `list_profiles()` hide otherwise healthy profiles. The fix reuses the canonical `iter_skill_index_files()` walker used by skill discovery and guards unexpected `OSError`s while preserving the existing count cache.

## Related Issue

Fixes #97173

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `hermes_cli/profiles.py`
  - use `agent.skill_utils.iter_skill_index_files()` instead of direct `Path.rglob()` for profile skill counts
  - return a degraded count instead of aborting the roster when root/scan operations raise `OSError`
  - preserve the existing signature/TTL cache
- `tests/hermes_cli/test_profiles.py`
  - add a regression test that raises `OSError(448, ...)` from `os.scandir()` for one child subtree
  - prove the profile remains visible and its readable skill is still counted

## How to Test

1. Run the targeted profile-list tests:
   ```bash
   scripts/run_tests.sh tests/hermes_cli/test_profiles.py -k TestListProfiles -q
   ```
   Result: **3 passed**.
2. Run the canonical walker tests:
   ```bash
   pytest tests/agent/test_skill_utils.py -q
   ```
   Result: **22 passed, 1 skipped**.
3. Run lint and syntax checks:
   ```bash
   ruff check hermes_cli/profiles.py tests/hermes_cli/test_profiles.py
   python -m py_compile hermes_cli/profiles.py tests/hermes_cli/test_profiles.py
   git diff --check
   ```
   Result: **passed**.
4. Regression-test sabotage check: restoring the exact `origin/main` `_count_skills()` implementation makes the new test fail with `OSError: [Errno 448] untrusted mount point`; restoring this fix makes it pass.

Windows baseline note: `tests/hermes_cli/test_profiles.py` has the same 6 pre-existing POSIX-assumption failures on both current `origin/main` and this branch. The branch adds one passing test (`51 passed, 6 failed`) versus the baseline (`50 passed, 6 failed`); no new failure was introduced.

### Code

- [x] I've tested on my platform: Windows 11, Python 3.11.15

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A; behavior-only bug fix with regression coverage

## Screenshots / Logs

Pre-fix regression failure excerpt:

```text
OSError: [Errno 448] untrusted mount point: '.../skills/blocked-junction'
hermes_cli/profiles.py:857: in _count_skills
    for md in skills_dir.rglob("SKILL.md"):
```

Post-fix targeted result:

```text
3 tests passed, 0 failed
```

### DIFF
--- a/hermes_cli/profiles.py
+++ b/hermes_cli/profiles.py
@@ -33,7 +33,7 @@ from dataclasses import dataclass
 from pathlib import Path, PurePosixPath, PureWindowsPath
 from typing import Dict, List, Optional, Tuple
 
-from agent.skill_utils import is_excluded_skill_path
+from agent.skill_utils import iter_skill_index_files
 from hermes_constants import (
     clear_named_profile_deleted,
     mark_named_profile_deleted,
@@ -839,7 +839,11 @@ def _skills_dir_signature(skills_dir: Path) -> float:
 def _count_skills(profile_dir: Path) -> int:
     """Count installed skills in a profile (cached by skills-dir signature)."""
     skills_dir = profile_dir / "skills"
-    if not skills_dir.is_dir():
+    try:
+        skills_dir_exists = skills_dir.is_dir()
+    except OSError:
+        return 0
+    if not skills_dir_exists:
         return 0
 
     key = str(skills_dir)
@@ -853,11 +857,10 @@ def _count_skills(profile_dir: Path) -> int:
     ):
         return cached[2]
 
-    count = 0
-    for md in skills_dir.rglob("SKILL.md"):
-        if is_excluded_skill_path(md):
-            continue
-        count += 1
+    try:
+        count = sum(1 for _ in iter_skill_index_files(skills_dir, "SKILL.md"))
+    except OSError:
+        count = 0
     _SKILL_COUNT_CACHE[key] = (signature, now, count)
     return count
 
--- a/tests/hermes_cli/test_profiles.py
+++ b/tests/hermes_cli/test_profiles.py
@@ -489,6 +489,30 @@ class TestListProfiles:
         assert "alpha" in names
         assert "beta" in names
 
+    def test_unreadable_skill_tree_does_not_hide_profile(
+        self, profile_env, monkeypatch,
+    ):
+        create_profile("blocked", no_alias=True)
+        skills_dir = get_profile_dir("blocked") / "skills"
+        readable_skill = skills_dir / "readable"
+        readable_skill.mkdir()
+        (readable_skill / "SKILL.md").write_text("# Readable", encoding="utf-8")
+        blocked_tree = skills_dir / "blocked-junction"
+        blocked_tree.mkdir()
+
+        real_scandir = os.scandir
+
+        def scandir_with_blocked_tree(path):
+            if os.fspath(path) == os.fspath(blocked_tree):
+                raise OSError(448, "untrusted mount point", str(path))
+            return real_scandir(path)
+
+        monkeypatch.setattr(os, "scandir", scandir_with_blocked_tree)
+
+        listed = {profile.name: profile for profile in list_profiles()}
+        assert "blocked" in listed
+        assert listed["blocked"].skill_count == 1
+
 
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Makes profile skill counting resilient to unreadable skill trees.

**What it does**
- `hermes_cli/profiles.py` switches from `is_excluded_skill_path` to `iter_skill_index_files` for counting, and wraps `_count_skills` / `_skills_dir_signature` to tolerate `PermissionError`/`OSError` so an unreadable `skills/` subtree doesn't hide the profile from `list`/`get_profile_dir`.
- `tests/hermes_cli/test_profiles.py` adds `test_unreadable_skill_tree_does_not_hide_profile` with a mocked unreadable dir.

**Non-blocking notes**
- `iter_skill_index_files` may silently skip unreadable files — consider debug log when a skill dir is skipped due to permission error, to help user diagnose why a skill isn't counted.
- Import change from `is_excluded_skill_path` to `iter_skill_index_files` — verify no remaining callsite still expects old helper; grep shows only count/signature paths use it, so fine.

Good defensive fix for elevated-installer scenarios (see also #97212).

PR #92338, opened 2026-08-22
feat(i18n): localize remaining gateway static bubbles

vesc-baek:
## Summary

`display.language` already translates approval prompts and a handful of gateway slash replies. The **same class** of user-facing copy was still hardcoded English:

- busy-input acknowledgements (`⏩ Steered into current run…`, queue, interrupt, redirect)
- first-time onboarding tips (gateway + CLI)
- `/queue` (gateway, Discord slash, ACP, CLI preview)
- `⏳ Working — N min` heartbeat
- session-failure notices (storage, context overflow, empty/interrupted turn)

This wires those strings through `t()` and adds catalog keys for every `SUPPORTED_LANGUAGES` locale.

A thinner busy-ack-only PR (#92301) was closed as spam-shaped. This is that whole remaining static surface in one change.

**Still English on purpose** (existing i18n scope): agent replies, tool output, logs, error tracebacks, slash-command descriptions.

## Test plan

- [x] `./scripts/run_tests.sh tests/agent/test_i18n.py tests/gateway/test_busy_session_ack.py tests/gateway/test_subagent_protection_30170.py tests/gateway/test_normalize_empty_agent_response.py tests/gateway/test_run_progress_interrupt.py` (64 passed)
- [ ] `display.language: ko` → busy Slack/Telegram session shows the Korean steer/queue/interrupt bubble
- [ ] Unset / `en` → existing English strings unchanged
- [ ] Generic heartbeat mode still uses `status_phrases` (not this catalog)

PR #97179, opened 2026-08-28
fix(i18n): route gateway/provider status notices through the locale catalog

FanyinLiu:
## What does this PR do?

Dozens of user-facing status notices were hardcoded English in `gateway/run.py`, `agent/conversation_loop.py`, `run_agent.py`, and `agent/chat_completion_helpers.py`:

- provider wait notices ("waiting on {model} — Ns with no response/output yet", auto-reconnect suffix)
- gateway restart / draining / queued-after-return states
- context-compression timeout and commit-overrun warnings
- API-call activity detail (call start/done, retry backoff strings)
- `/queue` and `/steer` usage + error replies, `/goal` `/loop` busy replies
- quick-command errors, hook-blocked, /learn /init /suggestions failure notices

This routes all of them through `agent.i18n.t()` and adds the `status.*` keys to **every shipped locale** (17 catalogs, fully translated — en/zh plus af/ar/de/es/fr/ga/hu/it/ja/ko/pt/ru/tr/uk/zh-hant).

**Scope notes:**
- Busy-input acks (steer/queue/interrupt/redirect acks, `⏳ Working` heartbeat) are deliberately **excluded** — open PR #92338 already routes that surface; this PR complements it without conflicting.
- Model-facing interrupt scaffold markers in `agent/conversation_loop.py` stay English per documented i18n scope (they exist for the model, not the user).
- In `agent/chat_completion_helpers.py` two local `t = threading.Thread(...)` variables were renamed to `worker` to un-shadow the conventional `t()` import.

No behavior change for English users: `en.yaml` values match the previous hardcoded strings verbatim.

## Related Issue

N/A — follow-up to the existing i18n effort (complements #92338).

## Type of Change

## Changes Made

- `gateway/run.py` — route restart/draining/queue-after-return + /queue /steer /goal /loop /quickcmd /hook /learn /init /suggestions notices through `t()`
- `agent/conversation_loop.py` — compression timeout/commit-overrun warnings + API-call activity detail through `t()`
- `agent/chat_completion_helpers.py` — provider wait notices + auto-reconnect suffix through `t()`; rename shadowing local `t` thread vars to `worker`
- `run_agent.py` — CLI-side status strings through `t()`
- `locales/*.yaml` — add 45 `status.*` keys to all 17 shipped catalogs

## How to Test

1. `pytest tests/agent/test_i18n.py -q` — catalog parity across all locales (36 passed)
2. `pytest tests/gateway/test_busy_session_ack.py tests/gateway/test_discord_free_response.py tests/gateway/test_subagent_protection_30170.py tests/agent/test_codex_ttfb_watchdog.py tests/agent/test_acp_provider_rails.py -q` — all pass
3. Set `display.language: zh`, run a gateway session, wait on a slow provider — the wait notice renders in Chinese; unset → identical English as before

### Code
- [x] I searched for existing PRs — complements #92338, no duplicate
- [x] I've run the affected test suites and they pass (full `tests/gateway`: only 6 pre-existing failures on untouched areas — scale_to_zero, systemd_notify, shutdown_forensics, weixin — reproduce identically on clean `origin/main`)
- [x] I've added tests for my changes — covered by existing catalog-parity tests (test_i18n.py asserts every locale has every key with matching placeholders)
- [x] I've tested on my platform: macOS 27 (Apple Silicon)

- [x] Docs updates — N/A (no config keys, no architecture change)
- [x] `cli-config.yaml.example` — N/A
- [x] `CONTRIBUTING.md`/`AGENTS.md` — N/A
- [x] Cross-platform — N/A (string routing only)

## Screenshots / Logs

```
tests/agent/test_i18n.py: 36 passed (catalog parity: keys + placeholders match English in all 17 locales)
```


Enough1122:

Solid i18n change. The `status:` block lands in every locale file with the full key set, the call sites consistently pass the same param names the templates expect, and the `t` → `worker` rename in `chat_completion_helpers.py` is necessary and correct — without it the local `t = threading.Thread(...)` would shadow the newly imported `t()` and break the very calls this PR adds. The catalog-parity test is the right safety net. A few things worth verifying before merge:

1. **`_touch_activity` strings are now localized — check consumers that substring-match them.** `agent.conversation_loop.py` localizes activity-touch strings ("starting API call #", "API call # completed", "tool results posted, continuing iteration #", "retry backoff", "error retry backoff", "empty response retry backoff", "API error recovery"). These feed the inactivity monitor and the session activity feed. If any consumer (gateway inactivity logic, TUI activity display, or a test) does `"retry backoff" in activity` or similar English substring matching, localization silently breaks the detection — worst case the inactivity monitor stops seeing liveness. Please grep for consumers of `_touch_activity`/`session_activity` that pattern-match these phrases.

2. **`_status_action_gerund()` now returns localized text — confirm nothing compares it to a literal.** The gerund ("restarting"/"shutting down", now localized) is interpolated into `queued_after_return`/`draining_another_turn`/`draining_new_work`. That's correct, but if any code elsewhere compares the gerund to the literal English value (e.g. `"restarting" in msg`, a status classifier, dedup) it will break for non-English locales. The diff only shows the three template consumers, but a quick grep for the literal strings would confirm no other reader exists.

3. **`status.queued` (with `{detail}`) — verify it has a consumer.** The code changes use `queued_next`, `queued_after_return`, `draining_another_turn`, `draining_new_work`, but I don't see a call site for the plain `queued` key. If the old "⏳ Queued for the next turn" messages moved to the `queued_*` variants, `queued` may be a dead catalog entry. If it's used elsewhere (a different surface), fine — worth confirming so an unused key doesn't ship in 15 locale files.

4. **The parity comment in `en.yaml` lists a stale subset.** It says "add it to EVERY locale file (en/zh/ja/de/es/fr/tr/uk)", but this PR adds the block to ~15 files (af, ar, de, en, es, fr, it, ja, ko, pt, ru, tr, uk, zh, zh-hant). Future contributors following that comment will think the list is 8 locales and may miss others. Update the comment to reflect the actual full set (or point at the parity test as the authority).

5. Minor: passing `exc=exc`/`e=e` where the value is an exception object renders via `str()` in `.format` — consistent with the previous `f"...: {exc}"`, so no regression. And the folded multi-line `hermes_working` strings across locales all render a real newline, matching en.yaml's `\n` — good.

Clean, well-scoped localization work; the two "verify no English-string consumer" points (#1, #2) are the ones with real regression potential.

FanyinLiu:
Thanks for the thorough pass — verified each point against the head commit:

1. **Activity-string consumers**: grepped the tree for runtime substring matching on the localized `_touch_activity` phrases ("retry backoff", "starting API call", "API call #", "tool results posted", "API error recovery") across gateway/, agent/, hermes_cli/, tui_gateway/ — no runtime consumer pattern-matches the text. The inactivity monitor keys off `_last_activity_ts` timestamps; the description is display/snapshot-only. The tests that reference the English literals set the values themselves, so nothing breaks under non-en locales.
2. **Gerund literal compares**: no code compares `_status_action_gerund()` output to a literal. The restart/shutdown handoff is driven by exit code 75 + the `resume_pending` session-store flag, not message text. (The two remaining hardcoded `action = "restarting"/"shutting down"` assignments in the shutdown/cron-interrupt notice paths are untouched by this PR and stay English — separate surface.)
3. **`status.queued` dead key**: confirmed — no call site. The busy-ack surface it was drafted for is owned by #92338 (which ships its own `queue`/`queue_subagent`/`queue_compression` keys), so this entry would ship dead in all 17 catalogs. Removed in 64870a8.
4. **Stale parity comment**: fixed in 64870a8 — the header now points at test_i18n.py as the parity authority instead of listing an 8-locale subset.
5. Ack — behavior unchanged from the previous f-string rendering.

`pytest tests/agent/test_i18n.py` (36 passed) and `tests/gateway/test_busy_session_ack.py` (10 passed) re-run green on the new head.

PR #97182, opened 2026-08-28
feat: expose terminal.temp_dir to redirect session temp root off tmpfs

rahlquist:
## Summary
Some Linux distros utilize RAM-based `tmpfs /tmp`  several Arch-based setups for example this is capped by RAM and  the temp directory is a small size. Hermes session temp files — background-process logs/pid/exit files and code-execution sandboxes — land in that temp root via `LocalEnvironment.get_temp_dir()`, so the agent runs out of space under load. Numerous times over long sessions I have had Hermes find this volume full. Blamelessly this may not be Hermes filling it but if we were able to re-point Hermes temp use elsewhere the point would be moot. 

This adds a first-class, documented `config.yaml` knob instead of forcing users to set `TMPDIR` process-wide:

- `hermes_cli/config_defaults.py`: new `terminal.temp_dir` key (default `""`).
- `hermes_cli/config.py`: map `temp_dir -> TERMINAL_TEMP_DIR` in `TERMINAL_CONFIG_ENV_MAP` so `apply_terminal_config_to_env` bridges it for child-process launch paths (TUI/dashboard/gateway workers) exactly like `cwd`.
- `tools/environments/local.py`: `get_temp_dir()` honors `TERMINAL_TEMP_DIR` first, falling through to `TMPDIR`/`/tmp`/`tempfile.gettempdir()` when unset or not an existing absolute dir. Windows still uses its `HERMES_HOME/cache/terminal` path.

Usage:
```yaml
terminal:
  temp_dir: /var/hermes-tmp   # point session temp at real storage
```

Follow-up to Teknium's request for a reminder to make the temp dir configurable.

## Test plan
- Added `tests/tools/test_local_temp_dir.py` (4 cases: override honored, process-env honored, missing dir falls through, empty falls through to TMPDIR).
- `uv run --extra dev pytest tests/tools/test_local_temp_dir.py tests/tools/test_base_environment.py` → 31 passed.
- `ruff` lint clean on touched files.

## Risk
None for existing users: default is empty, so resolution is unchanged. Only an absolute, existing `terminal.temp_dir` redirects; invalid values fall back gracefully.

### DIFF
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -3598,6 +3598,7 @@ TERMINAL_CONFIG_ENV_MAP = {
     "modal_mode": "TERMINAL_MODAL_MODE",
     "degraded_mode": "TERMINAL_DEGRADED_MODE",
     "cwd": "TERMINAL_CWD",
+    "temp_dir": "TERMINAL_TEMP_DIR",
     "timeout": "TERMINAL_TIMEOUT",
     "lifetime_seconds": "TERMINAL_LIFETIME_SECONDS",
     "docker_image": "TERMINAL_DOCKER_IMAGE",
--- a/hermes_cli/config_defaults.py
+++ b/hermes_cli/config_defaults.py
@@ -382,6 +382,13 @@ DEFAULT_CONFIG = {
         # preserves the historical error + traceback behavior.
         "degraded_mode": "warn",
         "cwd": ".",  # Use current directory
+        # Root directory for Hermes' terminal session temp files (background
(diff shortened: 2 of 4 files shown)


teknium1:
Merged via PR #97205. Your commit was cherry-picked onto current main with your authorship preserved in git log. On top of your temp_dir knob we also changed the DEFAULT temp root to HERMES_HOME/cache/terminal (real storage, auto-pruned after 72h) per maintainer direction. Thanks!

Issue #42126, opened 2026-06-08
[Bug]: systemd `gateway run` exits 1 right after platforms connect (crash-loop) — regression after auto-update; runs fine standalone


EtherAura:
## Summary

After an auto-update (~2026-06-07), `hermes gateway run --replace` exits with
**status 1 about 1–2 seconds after all messaging platforms connect**, but **only
when launched by systemd**. The *identical* command run by hand (foreground) runs
indefinitely. With `Restart=always` this becomes a crash loop and the gateway is
unavailable. The pre-update build had run stably for weeks.

## Environment

- Host: Raspberry Pi 5, Gentoo Linux **aarch64**, Linux 6.12.x
- Hermes version: latest auto-update-channel build as of **2026-06-07 ~19:41** (it
  replaced a prior build that had run stably under the same unit for weeks; exact
  `hermes --version` string available on request)
- Service: **user-scope** systemd unit (`systemctl --user`), with linger enabled
  - `Type=simple`, `Restart=always`, `RestartSec=5`, `KillMode=mixed`
  - `ExecStart=…/venv/bin/python -m hermes_cli.main gateway run --replace`
  - `WorkingDirectory=$HERMES_HOME`, `Environment=HERMES_HOME=/home/hermes/.hermes`
- Platforms configured: email, telegram, homeassistant (+ a stale `feishu` entry)
- Provider: an OpenAI-compatible (Codex) chat provider. The exit happens at gateway
  startup, *before* any model inference, so it appears provider-independent.

## Symptom

Journal (with `StandardError=journal`):

```
⚕ Hermes Gateway Starting...
[Email] Connected as …
hermes-gateway.service: Main process exited, code=exited, status=1/FAILURE
hermes-gateway.service: Failed with result 'exit-code'.
Started Hermes Agent Gateway …            # Restart=always -> crash loop
```

- **No Python traceback** is emitted to the journal.
- `gateway_state.json` is written with all platforms `"state":"connected"`,
  `"gateway_state":"running"` (sometimes `"stopped"`), and **`"exit_reason": null`** —
  i.e. the process exits 1 without recording a reason.

## Key finding — it is strictly systemd-specific

The same command runs fine outside systemd:

```bash
# runs the full 70s, then timeout kills it (exit 124) — STABLE
timeout 70 env XDG_RUNTIME_DIR=/run/user/1001 \
[...]

PR #97022, opened 2026-08-28
fix(tui_gateway): route host_supervisor PID probe/kill through gateway.status (Windows os.kill(pid, 0) footgun)

LovePlayCode:
## What does this PR do?

Fixes a Windows correctness bug in the dashboard compute-host supervisor's PID liveness probe.

`host_supervisor._pid_alive` used `os.kill(pid, 0)` to ask "is this PID alive". On POSIX that signal-0 idiom is a genuine no-op permission check — but **on Windows it is not**. CPython maps `sig=0` to `CTRL_C_EVENT` (the two collide at the integer value 0) and routes it through `GenerateConsoleCtrlEvent(0, pid)`, which broadcasts Ctrl+C to the **entire console process group** containing the target PID ([bpo-14484](https://bugs.python.org/issue14484), open since 2012, won't be fixed for compat reasons).

So a read-only status check could kill the process it was inspecting, plus unrelated sibling processes sharing its console. This is precisely what `CONTRIBUTING.md` cross-platform rule 1 forbids and what `scripts/check-windows-footguns.py` scans for.

The fix routes both the probe and the kill through helpers the repo already maintains, rather than adding new ones:

- `_pid_alive` → `gateway.status._pid_exists(pid)` — psutil-backed, with a ctypes `OpenProcess` / `WaitForSingleObject` fallback on Windows, and zombie-aware (issue #42126). `CONTRIBUTING.md` already names this the preferred wrapper.
- `_terminate_pid` → `gateway.status.terminate_pid(pid)` / `terminate_pid(pid, force=True)` — does `taskkill /T /F` on Windows, so the force-kill actually cascades to children.

## Related Issue

Fixes #97019

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✅ Tests (adding or improving test coverage)

## Changes Made

- **`tui_gateway/host_supervisor.py`**
  - `_pid_alive`: replace `os.kill(pid, 0)` with `gateway.status._pid_exists`. Also drops the old `PermissionError → True` branch, which conflated "exists but owned by another session" with an unconditional alive; the canonical probe distinguishes these.
  - `_terminate_pid`: replace `os.kill(pid, SIGTERM)` / `os.kill(pid, signal.SIGKILL)` with `gateway.status.terminate_pid`. `signal.SIGKILL` does not exist on Windows (`AttributeError` at import time) and bare `SIGTERM` there does not cascade to child processes.
  - Removes the now-unused `import signal`.
[...]

PR #94352, opened 2026-08-25
fix(tui): stop Windows pid probe from killing the host

Sahilvishnaliya:
## Problem

`_pid_alive()` in `tui_gateway/host_supervisor.py:88` probed liveness with bare `os.kill(pid, 0)` — the POSIX idiom. On Windows, sig=0 is **not** a no-op (bpo-14484): CPython maps it through `GenerateConsoleCtrlEvent`/`TerminateProcess`, so every "is this host still alive?" check **terminated the healthy TUI host process it was probing**, often along with siblings in the same console group.

Callers affected: registry staleness checks (`:226`) and the shutdown path (`:543`). This was the last runtime-reachable unguarded `os.kill(pid, 0)` in the repo — all others are annotated `windows-footgun: ok` with explicit win32 guards, and `scripts/check-windows-footguns.py` exists specifically to police this pattern.

## Fix

Delegate to the canonical cross-platform probe `gateway.status._pid_exists` (psutil-first with zombie handling, Win32 `OpenProcess`/`WaitForSingleObject` fallback, POSIX-only `os.kill` branch) — the same helper the gateway uses for exactly this reason. Function-level import keeps tui_gateway startup light; no import cycle (other `tui_gateway` modules already import from `gateway.*`, and `gateway.status` imports only stdlib + `hermes_constants` + `utils`).

The `pid <= 0` early return is preserved (POSIX `os.kill(0, 0)` would signal the whole process group).

## Verification

New `tests/tui_gateway/test_host_supervisor_pid_alive.py` (3 tests):
- live PID (`os.getpid()`) probes `True` with `os.kill` monkeypatched to raise — proves no signal path on any platform
- non-positive PIDs return `False` without touching signals
- a fully reaped child PID reports dead

3/3 pass; `scripts/check-windows-footguns.py --all` clean (1010 files).

PR #68272, opened 2026-07-20
fix(tui_gateway): make compute-host orphan cleanup Windows-safe

brainx:
## What does this PR do?

Makes dashboard compute-host orphan reconciliation safe on native Windows.

The supervisor used three POSIX assumptions in a cross-platform path: `os.kill(pid, 0)` for liveness, `/proc`/`ps` for process identity, and bare `signal.SIGKILL` for escalation. On Windows, the liveness probe can deliver Ctrl+C to the target console group, `/proc` and `ps` are unavailable, and `SIGKILL` does not exist.

This change delegates those operations to the existing cross-platform helpers in `gateway.status`. Command-line inspection remains fail-closed, so an unreadable or mismatched PID is never terminated.

## Related Issue

Related: #41662 and #68031. Those cover the same Windows signal-zero defect class in other components; neither changes `tui_gateway/host_supervisor.py`.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Route compute-host PID liveness through the psutil-backed, zombie-aware `_pid_exists()` helper.
- Route process command-line inspection through the existing Windows-aware `_read_process_cmdline()` helper.
- Route graceful and forced termination through `terminate_pid()`, including Windows `taskkill /T /F` escalation.
- Add `tui_gateway/` to the full Windows-footgun scan.
- Add regressions for safe liveness probing, Windows command-line identity, and cross-platform termination.

## How to Test

1. `scripts/run_tests.sh tests/tui_gateway/test_compute_host_phase1.py -q` — 17 passed.
2. `scripts/run_tests.sh tests/gateway/test_status.py -q -k 'TerminatePid or ReadProcessCmdline'` — 7 passed.
3. `.venv/bin/python scripts/check-windows-footguns.py --all` — no findings across 847 files.
4. `.venv/bin/ruff check tui_gateway/host_supervisor.py tests/tui_gateway/test_compute_host_phase1.py scripts/check-windows-footguns.py` — passed.

The three new regressions were also run before the implementation and failed on the unsafe liveness, identity, and termination paths as expected.

### Code

[...]

PR #97187, opened 2026-08-28
fix(tui-gateway): stop using os.kill(pid, 0) as liveness probe on Windows (#97019)

### LINKED ISSUE #97019: [Bug]: host_supervisor uses os.kill(pid, 0) for liveness, unsafe on Windows
## Bug Description

`tui_gateway/host_supervisor.py::\_pid_alive` uses `os.kill(pid, 0)` as a liveness probe. On Windows this is **not** a no-op: CPython maps `sig=0` to `CTRL_C_EVENT` (the two collide at the integer value 0) and routes it through `GenerateConsoleCtrlEvent(0, pid)`, which broadcasts Ctrl+C to the **entire console process group** containing the target PID. See [bpo-14484](https://bugs.python.org/issue14484) (open since 2012).

So a read-only status check can kill the process it is inspecting, plus unrelated sibling processes sharing its console.

This is the exact pattern `CONTRIBUTING.md` rule 1 forbids:

> **Never call `os.kill(pid, 0)` for liveness checks.**

and the one `scripts/check-windows-footguns.py` scans for.

The relevant caller is `HostSupervisor.reconcile_startup_orphan` (`tui_gateway/host_supervisor.py:231`) — a startup path that only wants to know whether a previously registered compute host is still running:

```python
if pid <= 0 or not _pid_alive(pid):
    self._remove_registry()
    return "not-running"
```

## Steps to Reproduce

1. On native Windows, start Hermes with `dashboard.turn_isolation` enabled so a `python -m tui_gateway.compute_host` child is spawned and its PID is written to `~/.hermes/state/dashboard-compute-host.json`.
2. Let the dashboard process exit without cleanly reaping the child (crash, or `CTRL_CLOSE_EVENT`), leaving a valid registry entry pointing at a live PID.
3. Restart Hermes so `reconcile_startup_orphan()` runs against that registry entry.
4. The probe fires on the still-live compute host PID.

## Expected Behavior

`_pid_alive` is a read-only query. It must never be able to terminate the process it inspects, nor any process sharing its console group.

## Actual Behavior

The probe sends `CTRL_C_EVENT` to the target's console process group. Depending on group membership this kills the compute host and potentially unrelated console-sharing processes.

## Affected Component

- Other (`tui_gateway` / dashboard compute-host supervisor)

## Operating System

Windows 11 (native). The bug is Windows-specific by nature; POSIX `os.kill(pid, 0)` is a genuine no-op permission check and behaves correctly.

## Root Cause Analysis

`tui_gateway/host_supervisor.py:87-104`:

```python
def _pid_alive(pid: int) -> bool:
    if pid <= 0:
        return False
    try:
        os.kill(pid, 0)
        return True
    except ProcessLookupError:
        return False
    except PermissionError:
        return True
```

Two problems:

1. **The probe can kill.** `os.kill(pid, 0)` on Windows is `GenerateConsoleCtrlEvent(0, pid)`.
2. **The `PermissionError` branch is wrong.** It returns `True` (alive) on access-denied, conflating "exists but owned by another user/session" with an unconditional alive. The canonical probe distinguishes these.

The repo already has the correct implementation — `gateway/status.py:821 _pid_exists()` — which is psutil-backed with a ctypes `OpenProcess` / `WaitForSingleObject` fallback on Windows, and additionally handles zombie processes (issue #42126). `CONTRIBUTING.md` names it as the preferred wrapper. This call site simply predates/converges on it.

A secondary issue in the same function: `_terminate_pid` (`:538`) calls `os.kill(pid, signal.SIGKILL)` on the escalation path. `signal.SIGKILL` does not exist on Windows — `AttributeError` at import time — and bare `SIGTERM` there does not cascade to child processes. `gateway.status.terminate_pid(force=True)` already implements the correct `taskkill /T /F` behavior.

## Proposed Fix

Route both paths through `gateway.status`:

- `_pid_alive` → `gateway.status._pid_exists(pid)`
- `_terminate_pid` → `gateway.status.terminate_pid(pid)` / `terminate_pid(pid, force=True)`

Note a caveat worth deciding during review: `_pid_alive` currently swallows *import* failures of `gateway.status` as `False` ("dead"), which is indistinguishable from "the process is gone". If that import ever fails (e.g. a missing transitive dep — `gateway/__init__.py` pulls in `hermes_cli.config`, which imports `yaml`), `reconcile_startup_orphan` concludes the old host died, removes the registry, and spawns a second compute host alongside the live one. Separating "probe unavailable" from "process dead" (or at least raising the log level above `debug`) would make that failure visible rather than silent.

I have a branch with the core fix + regression tests ready and will link the PR.

sanjaynandanj:
## Summary

Closes #97019.

\`tui_gateway/host_supervisor._pid_alive\` used \`os.kill(pid, 0)\` as a read-only liveness probe. On Windows this is **not** a no-op — CPython maps \`sig=0\` to \`CTRL_C_EVENT\` (they collide at the integer value 0) and routes it through \`GenerateConsoleCtrlEvent(0, pid)\`, broadcasting Ctrl+C to the entire console process group containing the target PID. So a read-only status check from \`reconcile_startup_orphan\` could kill the compute host it was inspecting — and unrelated sibling processes sharing its console. This is the exact pattern \`CONTRIBUTING.md\` rule 1 forbids and that \`scripts/check-windows-footguns.py\` scans for (bpo-14484).

The repo already has the correct probe at \`gateway/status._pid_exists\` (psutil-backed with a ctypes \`OpenProcess\` / \`WaitForSingleObject\` fallback, also handles zombie processes per issue #42126). This PR routes \`_pid_alive\` through it.

While in the file: \`_terminate_pid\` had the sibling bug — it called \`os.kill(pid, signal.SIGKILL)\` on the escalation path. \`signal.SIGKILL\` does not exist on Windows and a bare \`SIGTERM\` there does not cascade to child processes. Now routes through \`gateway.status.terminate_pid(pid, force=True)\` which uses \`taskkill /T /F\`.

Also addressed the caveat from the issue: if the \`gateway.status\` import fails (e.g. a missing transitive dep), the probe now logs a warning instead of silently returning \`False\`, so \"probe unavailable\" is not indistinguishable from \"process dead\" — otherwise \`reconcile_startup_orphan\` could remove the registry entry and spawn a second compute host next to the live one.

## Test plan

- [ ] On Windows: start Hermes with \`dashboard.turn_isolation\` enabled, kill the dashboard uncleanly, restart Hermes. Verify \`reconcile_startup_orphan\` correctly detects the still-live compute host and does not send Ctrl+C to its console group.
- [ ] On Windows: force the shutdown escalation path (\`_terminate_pid\` with an unresponsive child). Verify \`taskkill /T /F\` fires and the process tree exits, without \`AttributeError: SIGKILL\`.
- [ ] On POSIX: existing behaviour unchanged — \`_pid_exists\` uses psutil / \`os.kill(pid, 0)\` internally, and \`terminate_pid\` uses SIGTERM/SIGKILL.
- [ ] Confirm existing host_supervisor tests still pass.

### DIFF
--- a/tui_gateway/host_supervisor.py
+++ b/tui_gateway/host_supervisor.py
@@ -20,7 +20,7 @@ import time
 import uuid
 from collections.abc import Callable
 from pathlib import Path
-from typing import Any
+from typing import Any, Optional
 
 from hermes_constants import get_hermes_home
 from tools.environments.local import hermes_subprocess_env
@@ -85,37 +85,49 @@ def _default_registry_path() -> Path:
     return get_hermes_home() / "state" / _REGISTRY_NAME
 
 
-def _pid_alive(pid: int) -> bool:
-    # Never call os.kill(pid, 0) directly: on Windows CPython maps sig=0 to
-    # CTRL_C_EVENT (they collide at the C level) and routes it through
-    # GenerateConsoleCtrlEvent(0, pid), broadcasting Ctrl+C to the entire
-    # console process group containing the target PID. A read-only status
-    # check would then kill the process it inspects — and sibling processes
-    # sharing its console. See bpo-14484 and CONTRIBUTING.md rule 1.
-    # gateway.status._pid_exists is the canonical psutil-backed probe (with
-    # a ctypes OpenProcess/WaitForSingleObject fallback on Windows) and also
-    # handles zombie processes (issue #42126).
+_pid_probe_unavailable_logged = False
+
+
+def _pid_alive(pid: int) -> Optional[bool]:
+    """Return True if ``pid`` is alive, False if definitely dead, None if unknown.
+
+    Callers must treat ``None`` as "probe unavailable" and fail toward the
+    safer choice for their context — e.g. ``reconcile_startup_orphan`` must
+    not remove the registry on ``None`` or it can spawn a duplicate compute
+    host next to the live one.
+
+    Never call ``os.kill(pid, 0)`` directly: on Windows CPython maps sig=0 to
+    CTRL_C_EVENT (they collide at the C level) and routes it through
+    ``GenerateConsoleCtrlEvent(0, pid)``, broadcasting Ctrl+C to the entire
+    console process group containing the target PID. A read-only status
+    check would then kill the process it inspects — and sibling processes
+    sharing its console. See bpo-14484 and CONTRIBUTING.md rule 1.
+    ``gateway.status._pid_exists`` is the canonical psutil-backed probe
+    (with a ctypes ``OpenProcess``/``WaitForSingleObject`` fallback on
+    Windows) and also handles zombie processes (issue #42126).
+    """
+    global _pid_probe_unavailable_logged
     if pid <= 0:
         return False
     try:
         from gateway.status import _pid_exists
     except Exception:
-        # Log louder than debug so a missing probe (import-time failure of a
-        # transitive dep) is not silently indistinguishable from "process
-        # dead" — otherwise reconcile_startup_orphan concludes the old host
-        # died, removes the registry, and can spawn a second compute host
-        # alongside the live one.
-        logger.warning(
-            "gateway.status._pid_exists unavailable — treating pid=%s as gone",
-            pid,
-            exc_info=True,
-        )
-        return False
+        # Log once (not per-call): _pid_alive is invoked in a 50ms polling
+        # loop inside _terminate_pid, so a per-call warning with exc_info
+        # would emit a stack trace every 50ms if the import stays broken.
+        if not _pid_probe_unavailable_logged:
+            _pid_probe_unavailable_logged = True
+            logger.warning(
+                "gateway.status._pid_exists unavailable — pid liveness probe "
+                "will return unknown; callers must not treat as dead",
+                exc_info=True,
+            )
+        return None
     try:
         return bool(_pid_exists(int(pid)))
     except Exception:
         logger.debug("pid liveness probe failed for pid=%s", pid, exc_info=True)
-        return False
+        return None
 
 
 def _pid_command(pid: int) -> str:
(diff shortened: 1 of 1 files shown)


alt-glitch:

Related: #97022, #94352, and #68272 are open repairs for the same TUI compute-host liveness/termination path. This overlaps their mechanism; #97022 includes regression coverage and #68272 has broader supervisor scope. Maintainers should choose the preferred implementation.

Enough1122:

Good fix, and the reasoning is sound — `os.kill(pid, 0)` on Windows colliding with `CTRL_C_EVENT` and broadcasting Ctrl+C to the whole console group is a genuinely nasty failure mode (bpo-14484), and routing through the psutil-backed `_pid_exists` plus `terminate_pid(..., force=True)` (taskkill `/T /F` on Windows) is the right canonical path. The deferred imports are correct, and the zombie-handling change (`_pid_exists` treats zombies as dead where `os.kill(pid, 0)` treated them as alive) is an intentional improvement worth keeping.

A few points to consider:

1. **The probe-unavailable default points the wrong way for the duplicate-spawn risk you yourself documented.** When `gateway.status._pid_exists` can't be imported, the code logs WARNING and returns False (treat as *gone*). Your comment spells out the consequence: `reconcile_startup_orphan` then concludes the old host died, removes the registry, and can spawn a second compute host alongside the live one. For a liveness probe feeding an orphan-reconciliation/duplicate-prevention decision, the safe default when liveness is *unknown* is usually "treat as alive" (don't reclaim, don't re-spawn) — spawning a duplicate compute host is worse than briefly tolerating a stale registry entry. Consider returning a tri-state (alive / dead / unknown) or, at minimum, making the caller fail toward "don't spawn a second host" when the probe can't load. The current fail direction turns a transient import problem into a possible duplicate host.

2. **WARNING + `exc_info=True` inside a 0.05s polling loop will spam if the import keeps failing.** `_pid_alive` is called repeatedly from `_terminate_pid`'s wait loop (and possibly reconcile paths). If `gateway.status` has a persistent import issue, this is a stack trace every 50ms. Emit the loud warning once (module-level `warnings.warn`/a "logged once" guard) and keep the per-call path quiet — you still get the signal, but you don't flood logs while the loop burns through its deadline.

3. **Windows fallback force-kill degrades to a plain SIGTERM.** The fallback path's `sigkill = getattr(signal, "SIGKILL", signal.SIGTERM)` means that if `gateway.status` can't be imported *and* we're on Windows, the "force" escalation is just another SIGTERM that won't cascade to children. The comment says this path "should not fire in practice", and the primary route is correct — but if it ever does fire on Windows, the escalation silently does nothing. A warning there (or removing the os.kill fallback entirely and surfacing the missing dependency) would make the degradation visible instead of silent.

4. Minor: consider a one-line docstring on `_pid_alive` (it has a long comment but no docstring) stating the contract — "True = definitely alive, False = dead *or* probe unavailable" — since point 1 hinges on callers knowing that False is ambiguous.

Good, genuinely correct improvement — the fail-direction and logging-spam points are the only things I'd adjust.

PR #97188, opened 2026-08-28
feat(stt): honor stt.mistral.base_url, completing endpoint parity

meltforce:
## What does this PR do?

`stt.provider: mistral` can only ever address `api.mistral.ai`. `_transcribe_mistral` builds its client as `Mistral(api_key=api_key)` with no `server_url`, so no configuration reaches the endpoint.

Every other STT provider already honors a base URL from its own config block — `openai` (`_resolve_openai_stt_client_config`), `xai`, `elevenlabs` and `deepinfra`. Mistral is the exception.

The TTS half is not: #73512 introduced what its description calls *class-level `base_url` parity* for TTS providers, and `_generate_mistral_tts` has read `tts.mistral.base_url` into the SDK's `server_url` since. This is the same change for the transcription side, in the shape `_transcribe_elevenlabs` uses — config block, then env, then module constant.

Unset, no `server_url` is passed at all and the SDK keeps its own endpoint, so an install that does not configure this behaves exactly as before.

**Why it matters.** A self-hosted gateway or authenticating reverse proxy in front of Mistral is reachable for synthesis and unreachable for transcription, which makes the provider pair unusable in that deployment — with nothing in the configuration to indicate why. Concretely, this came up routing both directions through Tailscale Aperture, whose HTTP connector answers `/v1/audio/speech` and `/v1/audio/transcriptions` identically; TTS could be pointed at it, STT could not.

## Related Issue

None open that I could find — searched `stt base_url`, `voxtral`, and `voxtral endpoint` across issues and PRs per CONTRIBUTING § *Search First*. The closest prior art is #73512 (merged), which this continues.

## Type of Change

## Changes Made

- `tools/transcription_tools.py` — `_transcribe_mistral` resolves `stt.mistral.base_url` → `STT_MISTRAL_BASE_URL` → `MISTRAL_STT_BASE_URL` (new module constant, empty by default) and passes it to the SDK as `server_url` only when set.
- `tests/tools/test_transcription_tools.py` — three cases in `TestTranscribeMistral`: config `base_url` becomes `server_url` (with trailing-slash normalization), an unconfigured provider passes no `server_url` at all, and the env variable applies when the config block is silent.
- `website/docs/user-guide/features/tts.md` — the key in the STT config block, and a sentence in the Mistral provider detail.
- `website/docs/reference/environment-variables.md` — `STT_MISTRAL_BASE_URL`.

## How to Test

1. `python -m pytest tests/tools/test_transcription_tools.py -k "Mistral or mistral"` — 9 passed.
2. Regression across the transcription suite: `python -m pytest tests/tools/test_transcription_tools.py tests/tools/test_pre_transcription_hook.py tests/tools/test_transcription.py tests/tools/test_stt_language_resolution.py` — 101 passed.
3. End to end against a real proxy: set

   ```yaml
   stt:
     provider: mistral
     mistral:
       model: voxtral-mini-latest
       base_url: "https://<your-gateway>/v1/connectors/Mistral"
   ```

   and send a voice message. Verified this way on agent v0.20.6 against a Tailscale Aperture HTTP connector, which returned the correct transcript; without the change the same configuration reaches api.mistral.ai and the `base_url` value is ignored.

### DIFF
--- a/contributors/emails/25252086+meltforce@users.noreply.github.com
+++ b/contributors/emails/25252086+meltforce@users.noreply.github.com
@@ -1,0 +1,2 @@
+# stt.mistral.base_url endpoint parity
+meltforce
--- a/tests/tools/test_transcription_tools.py
+++ b/tests/tools/test_transcription_tools.py
@@ -625,6 +625,80 @@ class TestTranscribeMistral:
         assert "RuntimeError" in result["error"]
         assert "secret-key-leaked" not in result["error"]
 
+    def test_config_base_url_becomes_server_url(
+        self, monkeypatch, sample_ogg, mock_mistral_module
+    ):
+        """stt.mistral.base_url reaches the SDK as server_url.
+
+        Endpoint parity with tts.mistral.base_url and with every other STT
+        provider (openai, xai, elevenlabs, deepinfra), which all read
+        base_url from their own config block.
+        """
+        monkeypatch.setenv("MISTRAL_API_KEY", "test-key")
+        mock_result = MagicMock()
+        mock_result.text = "routed"
+        mock_mistral_module.audio.transcriptions.complete.return_value = mock_result
+
+        from tools import transcription_tools
+        config = {"mistral": {"base_url": "http://localhost:4004/v1/proxy/"}}
+        with patch.object(transcription_tools, "_load_stt_config", return_value=config):
+            result = transcription_tools._transcribe_mistral(
+                sample_ogg, "voxtral-mini-latest"
+            )
+
+        assert result["success"] is True
+        import sys
+        kwargs = sys.modules["mistralai.client"].Mistral.call_args.kwargs
+        # Trailing slash stripped, same normalization the other providers apply.
+        assert kwargs["server_url"] == "http://localhost:4004/v1/proxy"
+        assert kwargs["api_key"] == "test-key"
+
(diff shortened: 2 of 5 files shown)


Enough1122:

**Overall:** Completes endpoint parity for Mistral STT — `stt.mistral.base_url` → SDK `server_url`.

**What it does**
- `tools/transcription_tools.py` adds `MISTRAL_STT_BASE_URL = os.getenv("STT_MISTRAL_BASE_URL", "")` (empty default keeps SDK default) and maps `config stt.mistral.base_url` to SDK `server_url` (mirrors `tts.mistral.base_url` pattern).
- `tests/tools/test_transcription_tools.py` adds `test_config_base_url_becomes_server_url` and docs `website/docs/...` updated with env var + config comment.

**Non-blocking notes**
- Env var is `STT_MISTRAL_BASE_URL` but config key is `stt.mistral.base_url` — mapping precedence is config > env > SDK default; verify code checks `config.get("base_url")` first then fallback to `MISTRAL_STT_BASE_URL`, not the reverse, so user config overrides env as documented for TTS.
- Test mocks `mistralai` module — ensure it asserts `server_url` kwarg name exactly as SDK expects (`server_url` vs `base_url`); current test checks `server_url`, correct.

Small parity completion; no breaking change when unset.

PR #97190, opened 2026-08-28
fix(browser): include managed Node in subprocess PATH

### LINKED ISSUE #97186: fix(browser): _build_browser_env() does not merge Hermes node dir into PATH - npx/agent-browser fails on Windows (real-profile launch broken)
## Bug Description

With `browser.use_real_profile: true`, the real-profile browser launch fails on Windows with:

```
browser.use_real_profile is on, but the real-profile browser failed to start: Der Befehl "node" ist entweder falsch geschrieben oder konnte nicht gefunden werden.
```

("The command "node" is either misspelled or could not be found.")

## Root Cause

`_find_agent_browser()` falls back to `npx agent-browser` (`tools/browser_use_cli.py` → argv `C:\...\hermes\node\npx.CMD --ignore-scripts --prefer-offline -y agent-browser@^0.26.0`). `npx.CMD` sets `NODE_EXE=%~dp0\node.exe` and falls back to bare `node` on PATH when `%~dp0` does not resolve (observed when launched via subprocess). The Hermes-managed node dir is NOT on the subprocess PATH because `_build_browser_env()` (`tools/browser_tool.py`) — unlike the other env builders (lines ~1294, ~3160, ~3542 which apply `_merge_browser_path`) — does not merge it:

```python
def _build_browser_env() -> dict:
    env = hermes_subprocess_env(inherit_credentials=False)
    for _key in _BROWSER_PASSTHROUGH_KEYS:
        if _key in os.environ:
            env[_key] = os.environ[_key]
    return env  # <-- PATH never merged with the Hermes node dir
```

## Fix (verified locally)

```python
    env["PATH"] = _merge_browser_path(env.get("PATH", ""))
    return env
```

With the Hermes node dir on PATH, `npx.CMD` resolves `node` and agent-browser launches (`agent-browser 0.26.0`).

## Environment

- OS: Windows 11 Pro (26200)
- hermes-agent 0.20.6, Hermes-managed node v22.22.3 (`AppData\Local\hermes\node`)
- Triggered via `browser_exec` with `browser.use_real_profile: true`

## Workaround

Prepend the Hermes node dir to PATH in the subprocess env (or export it in the parent process before starting Hermes).

MagMueller:
## Summary

- Add Hermes-managed Node directories to the credential-scrubbed environment used by every browser subprocess.
- Restore Windows real-profile launch when `npx.CMD` cannot resolve its sibling `node.exe`.
- Keep the existing browser credential allowlist and secret stripping unchanged.

Fixes #97186

## Tests

- `scripts/run_tests.sh tests/tools/test_browser_real_profile.py tests/tools/test_browser_homebrew_paths.py tests/tools/test_browser_npx_warmup.py tests/tools/test_browser_use_cli.py -q` (218 passed)
- `ruff check tools/browser_tool.py tests/tools/test_browser_real_profile.py`
- `git diff --check`

### DIFF
--- a/tests/tools/test_browser_real_profile.py
+++ b/tests/tools/test_browser_real_profile.py
@@ -167,6 +167,40 @@ class TestRealProfileCdpLaunch:
         assert cdp == "http://127.0.0.1:41000"
         self._reset()
 
+    def test_launch_env_includes_managed_node_dir(self, tmp_path):
+        import tools.browser_tool as bt
+        self._reset()
+        proc = Mock(returncode=0, stdout="", stderr="")
+        base_path = r"C:\Windows\System32"
+        managed_path = os.pathsep.join(
+            [r"C:\Users\test\AppData\Local\hermes\node", base_path]
+        )
+        captured = {}
+
+        def fake_run(_argv, **kwargs):
+            captured.update(kwargs)
+            return proc
+
+        with patch.object(bt, "_use_real_profile", return_value=True), \
+             patch("hermes_cli.browser_connect.detect_default_chromium", return_value="chrome"), \
+             patch("hermes_cli.browser_connect.snapshot_real_profile", return_value=(str(tmp_path), None)), \
+             patch.object(bt, "_agent_browser_get_cdp",
+                          side_effect=[None, "http://127.0.0.1:41000"]), \
+             patch.object(bt, "_find_agent_browser", return_value=r"C:\hermes\node\npx.CMD"), \
+             patch("tools.environments.local.hermes_subprocess_env",
+                   return_value={"PATH": base_path, "SCRUBBED": "1"}), \
+             patch.object(bt, "_merge_browser_path", return_value=managed_path) as merge_path, \
+             patch.object(bt.subprocess, "run", side_effect=fake_run), \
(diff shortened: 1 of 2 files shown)


andrexibiza (commented):
Exact-head review on `a9a7499f06f7e00176ee53103543bebadbe39c79`: the implementation matches #97186's live defect boundary. `_build_browser_env()` still starts from `hermes_subprocess_env(inherit_credentials=False)`, preserves the existing browser-only credential allowlist, and now applies the already-canonical `_merge_browser_path()` to that scrubbed PATH before any browser subprocess is launched. The added real-profile regression exercises the actual subprocess env handoff, verifies the managed Node PATH reaches `subprocess.run`, and keeps an unrelated scrubbed-env sentinel intact, so this is stronger than a helper-only assertion.

Landing edge is exact: current `main` is still the PR base `0abecf7a93948da386d3b23adba37475cc40fe53`; this PR is one submitted commit / two files and GitHub reports it mergeable. Exact-head hosted execution is green: CI `33180821859`, Docker `33180821143`, Nix `33180820824`. I found no remaining source or exact-object blocker in this carrier. #97198 is a duplicate implementation of the same one-line production change with weaker regression coverage; #97190 is the cleaner canonical carrier.

PR #96986, opened 2026-08-28
test(desktop): fix managed-ssh-update launcher stub path on macOS

dlgrv:
## Summary

One-line test fix: the POSIX managed-launcher test in `electron/managed-ssh-update.test.ts` stubs the remote updater with `hermesPath: '/bin/true'` — a Linux path. On macOS `true` lives at `/usr/bin/true`, so the stub is not found, the launcher publishes `rc=127`, and the test fails with `'127' !== '0'`.

The fix picks the stub path per platform, preserving the test's intent (a launcher binary that exits 0) on both platforms. No production code is touched.

- macOS (arm64): `npx vitest run --project electron electron/managed-ssh-update.test.ts` → **28/28 passed** (was 1 failed)
- Full electron project on macOS: **1934 passed | 6 skipped, 0 failed**
- Linux behavior unchanged (`/bin/true` still used there)

Closes #96975

PR #97194, opened 2026-08-28
test(desktop): pick platform-appropriate /bin/true stub in managed-ssh-update test (#96975)

### LINKED ISSUE #96975: test(desktop): managed-ssh-update POSIX launcher test fails on macOS (/bin/true does not exist)
## Bug Description

`electron/managed-ssh-update.test.ts` — test `POSIX managed launcher executes the updater command and atomically publishes its status` fails on macOS with exit code `127` (command not found), while passing on Linux CI.

## Root cause

The test stubs the remote launcher with `hermesPath: '/bin/true'` (line 291). On Linux `/bin/true` exists and returns 0, but on macOS the `true` binary lives at `/usr/bin/true` — `/bin/true` does not exist. The generated launcher runs `/bin/true update --yes` under `/bin/sh`, gets `rc=127`, and publishes `127` into the status file, so `assert.equal(status, '0')` fails.

(Side note: macOS also lacks `setsid`, so the launcher takes the `nohup` branch — that path works fine.)

## Steps to Reproduce

On macOS:

```bash
cd apps/desktop
npx vitest run --project electron electron/managed-ssh-update.test.ts
```

Expected: pass. Actual: `AssertionError: '127' !== '0'`.

## Suggested fix

Pick the stub path per platform, e.g.:

```ts
const truePath = process.platform === 'darwin' ? '/usr/bin/true' : '/bin/true'
...
hermesPath: truePath,
```

(Or `command -v true`-style resolution; either keeps the test's intent — a launcher binary that exits 0 — intact.)

## Environment

- macOS 26.5 (arm64), verified also on the reviewed head of PR #95781 — unrelated to that PR's changes (file untouched by it).
- Linux CI is unaffected (`/bin/true` exists), which is why this was not caught.

sanjaynandanj:
## Summary

Closes #96975.

\`electron/managed-ssh-update.test.ts::POSIX managed launcher executes the updater command and atomically publishes its status\` stubs the remote hermes binary with \`hermesPath: '/bin/true'\`. That file exists on Linux but not on macOS — the \`true\` binary lives at \`/usr/bin/true\` there, and \`/bin/true\` does not exist. The generated launcher runs \`/bin/true update --yes\` under \`/bin/sh\`, gets \`rc=127\`, publishes \`127\` into the status file, and \`assert.equal(status, '0')\` fails with \`AssertionError: '127' !== '0'\`. Linux CI is unaffected which is why this was not caught.

Pick the stub path per platform (\`/usr/bin/true\` on \`darwin\`, \`/bin/true\` elsewhere). Keeps the test's intent — a launcher binary that exits 0 — intact.

## Test plan

- [ ] On macOS: \`cd apps/desktop && npx vitest run --project electron electron/managed-ssh-update.test.ts\` — the previously-failing case now passes (status file contains \`0\`).
- [ ] On Linux: the same command still passes (Linux path unchanged).
- [ ] No other tests reference \`hermesPath\` in this file, so no cross-test impact.

### DIFF
--- a/apps/desktop/electron/managed-ssh-update.test.ts
+++ b/apps/desktop/electron/managed-ssh-update.test.ts
@@ -283,12 +283,17 @@ test('POSIX managed launcher is detached, correlation-scoped, and never publishe
 test('POSIX managed launcher executes the updater command and atomically publishes its status', async () => {
   const home = await mkdtemp(path.join(os.tmpdir(), 'hermes-managed-launch-'))
 
+  // `/bin/true` exists on Linux but not on macOS (its `true` lives at
+  // `/usr/bin/true`), so a hard-coded `/bin/true` stub makes the generated
+  // launcher exit with rc=127 under `/bin/sh` on macOS. Pick per platform.
+  const trueBin = process.platform === 'darwin' ? '/usr/bin/true' : '/bin/true'
+
   try {
     const command = buildPosixManagedUpdateLaunch(
       {
         ssh: { exec: async () => '' },
         platform: 'Linux',
-        hermesPath: '/bin/true',
+        hermesPath: trueBin,
         hermesHome: home
       },
       CORRELATION


alt-glitch:

Duplicate of #96986: both PRs make the same one-line Darwin test-fixture change from `/bin/true` to `/usr/bin/true`.

Enough1122:

Reasonable minimal fix for the macOS `rc=127` failure, and the comment correctly explains why. A couple of notes, proportionate to the change size:

1. **The ternary's "else" branch is `/bin/true`, which is not Windows-safe.** `process.platform === 'darwin' ? '/usr/bin/true' : '/bin/true'` means win32 also gets `/bin/true` — a path that doesn't exist on Windows. If this test can ever run on the Windows desktop lane (the file sits in the shared `electron/` test dir), it stays broken there. If the test is gated to POSIX hosts only (likely, given it's "POSIX managed launcher"), a one-line guard or comment stating that would make the boundary explicit.

2. **A host-agnostic stub would avoid OS-specific path knowledge entirely.** Instead of hardcoding `/bin/true` vs `/usr/bin/true`, create a tiny executable stub in a `mkdtemp` dir (`writeFile` + `chmod 0o755`, or a symlink to the host's real `true` resolved via `which`). That works on Linux, macOS, and Windows alike, and removes the "which OS has which true" maintenance from the test. The test already does `mkdtemp` for `home`, so a sibling stub dir is cheap.

3. **Minor semantic mix: `platform: 'Linux'` is hardcoded while `hermesPath` is now host-selected.** The launcher is generated for `platform: 'Linux'` but the stub path is chosen by the host platform — which only works because the generated launcher *executes on the host*. That's fine, but worth a comment (or deriving `platform` from the host too) so a reader isn't confused about why a "Linux" launcher runs against `/usr/bin/true` on macOS.

Small, correct change for its stated scope; the Windows fallback and a temp-stub alternative are the only things worth considering.

OutThisLife:
Superseded by #121997 — the cluster fix lands the same approach with the missing bug-class tests and, where noted in that PR, a canonicalization/reconcile fix the candidates missed. Your diagnosis and code are credited there (Co-authored-by / cherry-pick). Thanks for the report and the patch.

PR #97198, opened 2026-08-28
fix(browser): merge Hermes node dir into PATH in _build_browser_env (Closes #97186)

### LINKED ISSUE #97186: fix(browser): _build_browser_env() does not merge Hermes node dir into PATH - npx/agent-browser fails on Windows (real-profile launch broken)

Synxneuos:
## Summary
Fixes #97186

When launching an agent-browser subprocess or real-profile browser on Windows, `_build_browser_env()` creates a credential-scrubbed environment but previously omitted merging Hermes managed node directories into `PATH`. Consequently, `npx.CMD` cannot locate `node` when `%~dp0` fails to resolve in subprocess invocations.

## Changes
- In `tools/browser_tool.py`: updated `_build_browser_env()` to invoke `_merge_browser_path(env.get("PATH", ""))`, ensuring proper path resolution.
- Added test coverage in `tests/tools/test_browser_env_path.py`.

## Verification
- `pytest tests/tools/test_browser_env_path.py` passed (1 passed).

### DIFF
--- a/tests/tools/test_browser_env_path.py
+++ b/tests/tools/test_browser_env_path.py
@@ -1,13 +1,35 @@
 """Test that _build_browser_env merges Hermes node dir into PATH."""
 
+import os
 from unittest.mock import patch
-from tools.browser_tool import _build_browser_env
+from tools.browser_tool import _build_browser_env, _merge_browser_path
 
 
-def test_build_browser_env_merges_browser_path():
+def test_build_browser_env_merges_browser_path(tmp_path):
     """Verify _build_browser_env() merges Hermes node dirs into the subprocess PATH (Closes #97186)."""
-    with patch.dict("os.environ", {"SOME_TEST_KEY": "val"}, clear=False):
-        env = _build_browser_env()
-        assert "PATH" in env
-        assert isinstance(env["PATH"], str)
-        assert len(env["PATH"]) > 0
+    fake_node_bin = tmp_path / "hermes_node" / "bin"
+    fake_node_bin.mkdir(parents=True, exist_ok=True)
+    fake_bin_str = str(fake_node_bin)
+
+    with patch("tools.browser_tool._browser_candidate_path_dirs", return_value=[fake_bin_str]):
+        with patch.dict("os.environ", {"PATH": "existing_bin_path"}, clear=False):
+            env = _build_browser_env()
+            assert "PATH" in env
+            parts = env["PATH"].split(os.pathsep)
+            assert fake_bin_str in parts
+            assert parts[0] == fake_bin_str
+            assert "existing_bin_path" in parts
+
+
+def test_build_browser_env_path_deduplication(tmp_path):
+    """Verify _build_browser_env() is idempotent and does not duplicate existing PATH entries."""
+    fake_node_bin = tmp_path / "hermes_node" / "bin"
+    fake_node_bin.mkdir(parents=True, exist_ok=True)
+    fake_bin_str = str(fake_node_bin)
+
+    with patch("tools.browser_tool._browser_candidate_path_dirs", return_value=[fake_bin_str]):
+        initial_path = os.pathsep.join([fake_bin_str, "other_bin"])
+        with patch.dict("os.environ", {"PATH": initial_path}, clear=False):
+            env = _build_browser_env()
+            parts = env["PATH"].split(os.pathsep)
+            assert parts.count(fake_bin_str) == 1
--- a/tests/tools/test_browser_env_path.py
+++ b/tests/tools/test_browser_env_path.py
@@ -1,0 +1,13 @@
+"""Test that _build_browser_env merges Hermes node dir into PATH."""
+
+from unittest.mock import patch
+from tools.browser_tool import _build_browser_env
+
+
+def test_build_browser_env_merges_browser_path():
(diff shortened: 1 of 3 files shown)


alt-glitch:

Duplicate of #97190. It contains the same managed-Node PATH merge; #97190 is the earlier clean PR with stronger regression coverage.

Enough1122:

The change itself is reasonable — moving the node-dir merge into `_build_browser_env` so every caller (including the line 4879 spawn path) inherits the merged PATH, rather than only the two call sites that remembered to do it manually. The direction is right and `_merge_browser_path` already dedups existing entries so it's idempotent. Two issues to address:

1. **The test does not verify the fix — it would pass without the change.** `test_build_browser_env_merges_browser_path` only asserts `"PATH" in env`, `isinstance(env["PATH"], str)`, and `len(env["PATH"]) > 0`. A `_build_browser_env()` that never merged anything would pass all three. The regression it's meant to pin (#97186: browser subprocess missing the Hermes node dir) is exactly "PATH doesn't contain the node dir", so the test should assert the node dir path appears in `env["PATH"]` — e.g. call `_merge_browser_path("")`/the node-dir helper and assert its result is a prefix of the built env's PATH, or assert the specific Hermes node bin dir is in the pathsep-split list. As written it's a no-op test.

2. **The PR leaves redundant manual merges at the existing call sites.** Lines 1175 and 2575 already do `browser_env["PATH"] = _merge_browser_path(browser_env.get("PATH", ""))` right after calling `_build_browser_env()`. With the merge now inside `_build_browser_env`, those two call sites merge twice (harmless — `_merge_browser_path` dedups — but redundant). Either remove them now that the function owns the merge, or the PR should at least note they're no longer needed so a follow-up doesn't leave dead code. The whole point of consolidating into `_build_browser_env` was to stop scattering the merge at call sites; leaving two of them in place undercuts that.

Minor: `_merge_browser_path` only prepends dirs that pass `os.path.isdir`, so the "fix" is conditional on the Hermes node dir actually existing at env-build time. That's sensible defensive behavior, but worth a line in the PR description that if the node dir is created later, an env built before then won't get it — consistent with how the existing call sites already behaved, so not a regression.

Nice small consolidation; just make the test actually assert the node dir is present and clean up the now-redundant manual merges.

Synxneuos:
Thanks for the thorough review and suggestions @Enough1122!

Updated in commit `9acd423`:
1. **Strengthened Unit Tests**: Updated `tests/tools/test_browser_env_path.py` to directly verify that candidate node directories prepended by `_merge_browser_path` appear as prefixes in `env["PATH"]`, along with idempotency/deduplication tests.
2. **Removed Redundant Call Sites**: Cleaned up the redundant manual `_merge_browser_path` calls across `tools/browser_tool.py` (lines 1295, 3161, 3543) now that `_build_browser_env()` centralizes ownership of the merge for all subprocess spawn callers.

Enough1122:
Ack — verified at 9acd423: tests now assert the merged node dirs appear as PATH prefixes with dedup/idempotency checks, and the redundant manual `_merge_browser_path` calls at the previous call sites are removed now that `_build_browser_env()` owns the merge. Both points addressed, thanks.

Synxneuos:
All feedback addressed and verified:
- Consolidated `_merge_browser_path` into `_build_browser_env` and cleaned up redundant call sites.
- Added comprehensive unit tests in `tests/tools/test_browser_env_path.py` verifying node dir resolution & deduplication.
- Updated existing warmup mocks so all test suites run cleanly.

All CI checks are now **100% green** and passing. Ready for review and merge!

Issue #96531, opened 2026-08-27
[Bug]: Desktop app archive fails with "Session not found" when using non-default profile


EncoTime:
### Bug Description

The Desktop app does not pass the `profile` parameter when calling the archive API. As a result, the API queries the default profile's database instead of the active profile's database, returning "Session not found" for sessions that exist in a non-default profile.

### Steps to Reproduce

1. Connect the Desktop app to a non-default profile (e.g. `xiaomi` profile)
2. Open the session list and locate a session belonging to that profile
3. Click the archive button on that session
4. Observe the "Session not found" error

### Expected Behavior

The session should be archived successfully regardless of which profile is active.

### Actual Behavior

The archive operation fails with "Session not found" because the API defaults to querying the default profile's database instead of the active profile's database.

### Affected Component

Desktop app

### Messaging Platform (if gateway-related)

N/A (Desktop app issue)

### Debug Report

```
Session ID: 20260827_033649_b661bd
Active profile: xiaomi
Database path: /opt/data/profiles/xiaomi/state.db
Error: Session not found
Hermes backend: 0.20.6
Desktop frontend: 0.20.5
```

### Operating System

Linux (Docker container, Debian 13)

### Python Version

3.13.5

### Hermes Version

Backend: 0.20.6, Frontend: 0.20.5

### Additional Logs / Traceback (optional)

```
Relevant investigation log:
1. User reported inability to archive a session
2. No archive failure errors found in gateway logs
3. Session confirmed to exist in xiaomi profile's state.db
4. Desktop app is connected to xiaomi profile's gateway
5. Archive API call was made WITHOUT the profile parameter
6. API queried default profile's database → session not found
```

### Root Cause Analysis (optional)

The Desktop app's archive API call does not include the `profile` parameter. The server-side API defaults to the default profile's database when no profile is specified.

**Detailed flow:**
1. Session `20260827_033649_b661bd` exists in the `xiaomi` profile database (`/opt/data/profiles/xiaomi/state.db`)
2. Desktop app is connected to the `xiaomi` profile's gateway
3. Archive request is sent **without** a `profile` parameter
4. API resolves to the default profile's database, cannot find the session
5. Returns "Session not found"

**Impact:** All Desktop app users who use non-default profiles.

### Proposed Fix (optional)

1. Pass the current `profile` parameter in the Desktop app's archive API call
2. Audit all session-related API calls in the Desktop app to ensure they include the `profile` parameter
[...]

PR #44138, opened 2026-06-11
fix(desktop): pass owning profile to server on session delete/archive

AIalliAI:
Fixes #44117

## Problem

Deleting a session in the default profile from the desktop sidebar fails with **"Delete failed — Session not found"**, and the session stays in the list.

## Root cause

The desktop sidebar lists sessions via `GET /api/profiles/sessions`, which the primary backend serves by reading **every profile's `state.db` directly from disk** — so it always shows each profile's rows regardless of which profile the serving process itself runs as.

The mutations, however, are inconsistent about telling the *server* which profile owns the session:

| Operation | Profile reaches the server? |
|---|---|
| Read transcript (`getSessionMessages`) | ✅ `?profile=` query param |
| Rename (`renameSession`) | ✅ `profile` in PATCH body |
| **Delete (`deleteSession`)** | ❌ only `request.profile` (Electron's backend-process router) |
| **Archive (`setSessionArchived`)** | ❌ only `request.profile` |

When the dashboard process is actually scoped to a different profile than the desktop believes — e.g. after **"Set as active"** on the Profiles page, which flips the sticky `active_profile` file that the next legacy desktop launch (no `--profile` flag, see `_apply_profile_override` step 2) honors — `DELETE /api/sessions/{id}` opens the *serving process's own* `state.db`, doesn't find the row, and 404s. The same profile-confusion class was already fixed for the skills/toolsets endpoints (see the `_profile_scope` comment block in `hermes_cli/web_server.py`); session delete/archive were missed.

## Fix

Match the established read/rename pattern — pass the owning profile to the server, not just to Electron's process router:

- `deleteSession()`: append `?profile=` to the path (the endpoint already accepts it). `request.profile` is kept so per-profile remote overrides still route correctly; `interceptSessionRequestForRemote` reads the param from `searchParams` and global-remote mode re-appends it, so both remote shapes are unaffected.
- `setSessionArchived()`: include `profile` in the PATCH body, matching `renameSession()` (the endpoint's `SessionRename` model already reads `body.profile`).
- `delete_session_endpoint`: resolve the id via `resolve_session_id()` first, for parity with its GET/PATCH siblings (they all resolve; DELETE was the only one doing a raw exact match).

## Tests

- `tests/hermes_cli/test_web_server.py`: DELETE with `?profile=` deletes from that profile's `state.db` (regression test for this issue); DELETE resolves unique id prefixes; unknown id still 404s. Full file passes (265 passed).
[...]

PR #97199, opened 2026-08-28
fix(desktop): scope every session mutation to its owning profile

### LINKED ISSUE #78836: [Bug] Desktop app: deleting QQ/Telegram messaging sessions from sidebar flashes then reappears (DELETE routed to wrong profile backend)
## Bug Description

Deleting a **messaging-platform session** (QQBot / Telegram) from the **desktop app sidebar** does not actually delete the session — the row briefly disappears (optimistic removal) then comes right back on the next list refresh. Deleting a desktop-native session works fine.

## Steps to Reproduce

1. Have QQBot or Telegram gateway sessions in the desktop sidebar (source `qqbot` / `telegram`, belonging to a non-primary profile, e.g. `winefox` while the desktop primary backend is `default`).
2. Right-click a QQ/TG session row → click "Delete".
3. Observe: the row flashes away (optimistic UI removal) and then reappears within a second or two.

## Expected Behavior

The messaging-platform session should be permanently deleted from its owning profile's `state.db` and stay gone from the sidebar.

## Actual Behavior

The row reappears. The DELETE request does not reach the owning profile's backend, so nothing is actually deleted; the UI rollback/refresh brings the row back.

## Investigation / Likely Root Cause

The sidebar session list is loaded via the **cross-profile aggregator** endpoint:

- `GET /api/profiles/sessions/sidebar` (messaging slice uses `profile: 'all'`)

Session rows therefore include QQ/TG sessions owned by **non-primary profiles** (e.g. `winefox`), while the desktop's **primary backend** may be another profile (e.g. `default`).

The delete mutation in `apps/desktop/src/app/session/hooks/use-session-actions/index.ts` (`removeSession`) calls:

```ts
await deleteSession(storedSessionId, removed?.profile)
```

and `deleteSession` in `apps/desktop/src/hermes.ts` only forwards a `profile` when `removed?.profile` is truthy:

```ts
export function deleteSession(id: string, profile?: string | null) {
  return window.hermesDesktop.api({
    ...(profile ? { profile } : {}),   // <-- omitted when profile is null/undefined
    path: `/api/sessions/${encodeURIComponent(id)}`,
    method: 'DELETE'
  })
}
```

When the session object lacks a usable `profile` field (or it's `null`), the DELETE is routed to the **primary backend** (`ensureBackend(undefined)` in `electron/main.ts`), which opens the **wrong profile's** `state.db`. That DB has no row for the QQ/TG session, so the backend returns `{"ok": true, "already_absent": true}` — a **fake success**. The frontend treats it as success (optimistic removal), but the row still exists in the owning profile's DB, so the next list refresh resurrects it.

### Evidence collected on Windows (winefox profile)

- `winefox` backend (`127.0.0.1:8080`, spawned as `hermes --profile winefox serve`): `GET /api/sessions` lists 14 QQ/TG sessions; `DELETE /api/sessions/<id>` genuinely deletes one (verified row gone afterwards).
- `default` backend (`127.0.0.1:5910`): `GET /api/sessions` shows **0** QQ/TG sessions (total=0). A DELETE sent there for a QQ/TG id would hit `already_absent` and never touch the real row.
- The gateway being stopped/running is irrelevant — desktop session deletes go through the local `serve` backends, not the gateway.

## Environment

- OS: Windows 10
- Desktop app: packaged `Hermes.exe` (`apps/desktop/release/win-unpacked`)
- Hermes commit: `4075c8fd5ade673c93f757903b003ddc52603577`
- Multi-profile setup: desktop primary = `default`, QQ/TG sessions owned by `winefox` profile (profile name in `state.db` = `winefox`)
- Affected sources: `qqbot`, `telegram`

## Suggested Fix Direction

- Ensure the session rows returned by `/api/profiles/sessions/sidebar` carry the owning `profile` field, and/or
- In `removeSession`, fall back to resolving the session's owning profile (e.g. `resolveStoredSession`) when `removed?.profile` is missing before issuing the DELETE, so the request is routed to the correct backend.
- Consider having `DELETE /api/sessions/<id>` distinguish "already absent" from "deleted" so a mis-routed delete is visible instead of a fake success.

Bergmann89:
## What does this PR do?

Fixes a class of desktop bug where **session mutations don't stick** — deleting, archiving, pinning, or marking-unread a session from the sidebar appears to succeed (the row disappears optimistically) but silently no-ops, and the row reappears on the next list refresh.

The problem only shows up when the mutated session is **not** owned by the profile currently serving the desktop's backend — most visibly in the **"All Profiles"** view, and on a **remote-primary** desktop (a registered remote gateway is the primary and every profile is served from it).

Root cause: the four mutation helpers in `apps/desktop/src/api/sessions.ts` passed the owning profile **only** as `request.profile`. That field drives Electron's backend **routing**, but it is not the scope the backend reads to pick its target DB — the backend selects the DB from `?profile=` (DELETE) or `body.profile` (PATCH). Reads (`getSession`, `getSessionMessages`) and `renameSession` already scope correctly; the four mutations did not:

- `deleteSession` — no `?profile=` in the URL
- `setSessionArchived` — no `profile` in the PATCH body
- `setSessionPinnedRemote` — no `profile` in the PATCH body
- `setSessionUnreadRemote` — no `profile` in the PATCH body

On a remote gateway whose connection has no `remoteProfile` alias, the main process leaves such requests unscoped, so the backend opens its own (default) `state.db`, finds no matching row, and returns `{ok:true, already_absent:true}` (DELETE) or a silent no-op (PATCH) — a fake success the UI treats as done. `deleteSession` specifically lost its `?profile=` scope in the `src/hermes.ts` → `src/api/` module split (it was present via PR #44138 / the pre-split path).

The fix scopes all four mutations the same way the already-working endpoints do — the minimal, consistent approach that requires no backend or routing changes and leaves single-profile / selected-profile users byte-identical.

## Related Issue

Fixes #78836

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `apps/desktop/src/api/sessions.ts`
  - `deleteSession` — append `sessionScopeQuery(profile)` to the URL so the DELETE carries `?profile=<owner>` (mirrors `getSession` / `getSessionMessages`).
  - `setSessionArchived` — include `profile` in the PATCH body (mirrors `renameSession`).
  - `setSessionPinnedRemote` — include `profile` in the PATCH body.
  - `setSessionUnreadRemote` — include `profile` in the PATCH body.
  - `request.profile` is left in place on every mutation so per-profile remote-override and global-remote routing still resolve the correct backend.
- `apps/desktop/src/api/sessions.test.ts`
  - New coverage: `deleteSession` scopes `?profile=` in the URL for object and bare-string owners and omits it when no owner is known; `setSessionArchived` / `setSessionPinnedRemote` / `setSessionUnreadRemote` carry `body.profile` when owned and omit it otherwise.

Audited but unchanged (already correctly scoped or N/A): `listSessions`, `listAllProfileSessions`, `listSidebarSessions`, `getSession`, `getSessionMessages` (+ `getLatest`/`getOlder`/`getAll`), `renameSession`, `searchSessions`. The desktop frontend does not call the backend `bulk-delete` / `prune` / `empty` session endpoints, so those need no client change here.

## How to Test

Requires a multi-profile setup where the mutated session is owned by a **non-serving** profile — e.g. a remote-primary desktop (a registered remote gateway as primary), or the "All Profiles" view with sessions across profiles.

1. In the desktop sidebar, switch to the **"All Profiles on this Gateway"** view.
2. Delete a session that belongs to a non-primary profile (e.g. a `tommy` session while the serving/primary profile is `default`).
3. Switch profiles back and forth. **Before:** the row reappears (never deleted). **After:** the row stays gone. Repeat for archive, pin, and mark-unread — each now sticks.

Proof the fix reaches the right backend (remote-primary setup, gateway HTTPS access log):

```
# Before — no profile scope, backend hits the wrong (default) state.db:
"DELETE /api/sessions/<id> HTTP/1.1" 200 33      # {"ok":true,"already_absent":true}

# After — scoped to the owning profile, row actually removed:
"DELETE /api/sessions/<id>?profile=<owner> HTTP/1.1" 200 ...
```

Server-side confirmation: before the fix the "deleted" row was still present in the owning profile's `state.db`; after the fix it is gone.

Automated (from `apps/desktop`):

```
npx vitest run src/api/sessions.test.ts     # 10 passed
npx eslint src/api/sessions.ts src/api/sessions.test.ts   # clean
npx tsc -p . --noEmit                        # no new errors
```

### Code

- [ ] I've run `pytest tests/ -q` and all tests pass — N/A: desktop-only TypeScript change with no Python surface. Verified with the desktop suite instead: `vitest run src/api/sessions.test.ts` (10 passed), `eslint` clean, `tsc -p . --noEmit` clean.
- [x] I've tested on my platform: Ubuntu (X11), Electron 40 desktop build against a remote-primary Hermes gateway

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (internal API-helper scoping only; no doc-facing behavior change)
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — or N/A (no config keys)
- [x] I've updated `CONTRIBUTING.md` or `AGENTS.md` if I changed architecture or workflows — or N/A (no architecture change)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (platform-agnostic URL/body scoping; no OS-specific code)
- [x] I've updated tool descriptions/schemas if I changed tool behavior — or N/A (no tool behavior change)

## Screenshots / Logs

Gateway HTTPS reverse-proxy access log during the repro (remote-primary desktop):

```
Before (row silently survives):
[28/Aug/2026:13:01:55 +0000] "DELETE /api/sessions/20260828_082655_c8709d HTTP/1.1" 200 33 "Hermes/… Electron/40 …"

After (scoped, row deleted):
[.../…] "DELETE /api/sessions/20260828_082655_c8709d?profile=tommy HTTP/1.1" 200 …
```

The 33-byte body before the fix is `{"ok":true,"already_absent":true}` — the backend never found the row because it opened the wrong profile's `state.db`.

### DIFF
--- a/apps/desktop/src/api/sessions.test.ts
+++ b/apps/desktop/src/api/sessions.test.ts
@@ -10,13 +10,146 @@ vi.mock('./client', () => ({
 }))
 
 const client = await import('./client')
-const { listSidebarSessions } = await import('./sessions')
+
+const { deleteSession, setSessionArchived, setSessionPinnedRemote, setSessionUnreadRemote, listSidebarSessions } =
+  await import('./sessions')
 
 const hermesApi = vi.mocked(client.hermesApi)
 
 beforeEach(() => {
   vi.clearAllMocks()
   vi.mocked(client.getApiRequestConnection).mockReturnValue('prometheus')
+})
+
+describe('deleteSession profile scoping', () => {
+  it('scopes the DELETE to the owning profile in the URL (object owner)', async () => {
+    // Regression: the sidebar "All Profiles" delete sent the profile only via
+    // request.profile, not in the URL. On a remote gateway with no remoteProfile
+    // alias the main-process path rewrite left the URL unscoped, so the backend
+    // opened its own default state.db, missed the row, and returned
+    // {ok:true, already_absent:true} — the row vanished optimistically but was
+    // never deleted and came back on refresh. The URL must carry ?profile=.
+    hermesApi.mockResolvedValue({ ok: true } as never)
+    // Mirrors the real capabilityScoped for an object owner (remote-stamped row).
+    vi.mocked(client.capabilityScoped).mockReturnValue({ profile: 'tommy', connectionId: 'hermes-pi' })
+
+    await deleteSession('sess-1', { connectionId: 'hermes-pi', profile: 'tommy' })
+
+    expect(hermesApi.mock.calls[0][0]).toMatchObject({
+      method: 'DELETE',
+      path: '/api/sessions/sess-1?profile=tommy',
+      connectionId: 'hermes-pi',
+      profile: 'tommy'
+    })
+  })
+
+  it('scopes the DELETE to the owning profile in the URL (bare string owner)', async () => {
+    hermesApi.mockResolvedValue({ ok: true } as never)
+    // Bare-string owner: capabilityScoped resolves it to a profile scope.
+    vi.mocked(client.capabilityScoped).mockReturnValue({ profile: 'tommy' })
+
+    await deleteSession('sess-2', 'tommy')
+
+    expect(hermesApi.mock.calls[0][0]).toMatchObject({
+      method: 'DELETE',
+      path: '/api/sessions/sess-2?profile=tommy'
+    })
+  })
+
+  it('omits the profile query when no owner is known', async () => {
+    hermesApi.mockResolvedValue({ ok: true } as never)
+
+    await deleteSession('sess-3')
+
+    expect(hermesApi.mock.calls[0][0]).toMatchObject({
+      method: 'DELETE',
+      path: '/api/sessions/sess-3'
+    })
+    expect((hermesApi.mock.calls[0][0] as { path: string }).path).not.toContain('profile=')
+  })
+
+  it('keeps an explicit local pin routed to the local pool', async () => {
+    hermesApi.mockResolvedValue({ ok: true } as never)
+    // capabilityScoped drops a 'local' connection id by design; sessionScoped
+    // must re-add it so the request stays pinned to this device.
+    vi.mocked(client.capabilityScoped).mockReturnValue({ profile: 'tommy' })
+
+    await deleteSession('sess-4', { connectionId: 'local', profile: 'tommy' })
+
+    expect(hermesApi.mock.calls[0][0]).toMatchObject({
+      method: 'DELETE',
+      path: '/api/sessions/sess-4?profile=tommy',
+      connectionId: 'local',
+      profile: 'tommy'
+    })
+  })
+})
+
+describe('setSessionArchived profile scoping', () => {
+  it('carries the owning profile in the PATCH body', async () => {
+    // Same class as the unscoped DELETE: the PATCH handler reads its target DB
+    // from body.profile, so archiving a foreign-profile session must send it in
+    // the body, not only as request.profile (Electron routing), or on a remote
+    // gateway the archive lands on the wrong state.db and silently no-ops.
+    hermesApi.mockResolvedValue({ ok: true } as never)
+
+    await setSessionArchived('sess-a', true, 'tommy')
+
+    expect(hermesApi.mock.calls[0][0]).toMatchObject({
+      method: 'PATCH',
+      path: '/api/sessions/sess-a',
+      profile: 'tommy',
+      body: { archived: true, profile: 'tommy' }
+    })
+  })
+
+  it('omits the profile from the body when none is given', async () => {
+    hermesApi.mockResolvedValue({ ok: true } as never)
+
+    await setSessionArchived('sess-b', false)
+
+    const req = hermesApi.mock.calls[0][0] as { body: Record<string, unknown> }
+    expect(req).toMatchObject({ method: 'PATCH', body: { archived: false } })
+    expect(req.body).not.toHaveProperty('profile')
+  })
+})
+
+describe('setSessionPinnedRemote / setSessionUnreadRemote profile scoping', () => {
+  it('carries the owning profile in the pin PATCH body', async () => {
+    hermesApi.mockResolvedValue({ ok: true } as never)
+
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Scopes every desktop session mutation to its owning profile — fixes cross-profile session leakage.

**What it does**
- `apps/desktop/src/api/sessions.ts` carries owning `profile` in PATCH body for `setSessionArchived`, `setSessionPinnedRemote`, `setSessionUnreadRemote`, `deleteSession`, mirroring `renameSession` pattern; backend reads `profile` from body to route to correct remote/local scope. `listSidebarSessions` already profile-aware.
- `apps/desktop/src/api/sessions.test.ts` expands to cover profile routing for each mutation (134 lines).

**Non-blocking notes**
- PATCH body now includes `profile` — verify backend treats missing `profile` as "current/default" for backward compat with older desktop builds that don't send it; test mocks `getApiRequestConnection` returning `prometheus`, so default path still covered.
- No migration for in-flight sessions stuck in wrong profile after upgrade; a one-time reconciliation or doctor check could surface mis-scoped sessions, but out of scope.

Correct scoping; matches `renameSession` precedent.

teknium1:
Merged via #99376 (commit 5cc3da6827) — your commit was cherry-picked onto current main with your authorship preserved in git history (7c2ec7c73b on the salvage branch), since this branch was ~411 commits stale. Your diagnosis was exactly right: `request.profile` only drives Electron routing while the backend selects its target DB from `?profile=`/`body.profile`, and your fix covered all four mutations plus 134 lines of test coverage — it shipped essentially untouched, and resolved #96531, #98609, and canonical #78836. Thank you @Bergmann89, excellent work.

PR #97201, opened 2026-08-28
fix(kanban): reset tasks.started_at on every claim

sabastian-veroscale:
**Problem**

When a card is re-claimed (crash, stale claim, manual reclaim, review loop), `tasks.started_at` was NOT reset: `claim_task` and `claim_review_task` both used `COALESCE(started_at, ?)` on the claim UPDATE, so the column kept the original first-claim's timestamp.

Age-based consumers then measured **elapsed-since-first-claim** instead of the current worker's age:
- `task_age()` → `started_age_seconds` → the kanban dashboard's "age" field (plugins/kanban/dashboard/plugin_api.py)
- `detect_crashed_workers`' launch-window grace check (reads `tasks.started_at` directly) — a freshly spawned retry worker inherited a stale `started_at` from a previous claim, so the crash-grace window was already expired and the new worker could be flagged "pid not alive" before its PID was visible (spawn-race false positives).

**Fix**

Always overwrite `started_at` with the current claim's start time in both `claim_task` and `claim_review_task`. The field now tracks the ACTIVE worker's age. Per-attempt timing remains authoritative in `task_runs.started_at` — `enforce_max_runtime` / `detect_stale_running` already read `COALESCE(r.started_at, t.started_at)` and were correct.

**Tests**

- `test_reclaim_resets_started_at` — claim, release to ready, re-claim; asserts the second `started_at` is within seconds of the second claim and `task_age` reports ~0s.
- `test_reclaim_resets_started_at_review_loop` — review → changes → rework → re-review; asserts reviewer claim resets `started_at`.

Both fail on the old code (verified by reverting the fix locally and rerunning) and pass with it. Full kanban suites: 166 passed, 1 skipped.

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -4636,6 +4636,13 @@ def claim_task(
 
     Returns the claimed ``Task`` on success, ``None`` if the task was
     already claimed (or is not in ``ready`` status).
+
+    ``tasks.started_at`` is always overwritten with the current claim's
+    start time, so a re-claimed card (crash, stale claim, manual
+    reclaim) reflects the NEW worker's age rather than
+    elapsed-since-first-claim. Per-attempt timing lives in the
+    ``task_runs`` row; ``tasks.started_at`` is the active worker's
+    clock.
     """
     now = int(time.time())
     lock = claimer or _claimer_id()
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Correct claim-age fix so re-claimed kanban cards show new worker's age, not first-claim age.

**What it does**
- `hermes_cli/kanban_db.py:claim_task` now overwrites `tasks.started_at` with current claim time on every successful claim (`SET started_at = ?`) instead of `COALESCE(started_at, ?)` which preserved first claim's timestamp across re-claims (crash, stale, manual reclaim).
- Comment notes per-attempt timing lives in `task_runs` row; `tasks.started_at` is the active worker's age (used by `task_age`/dashboard).
- `tests/hermes_cli/test_kanban_db.py` adds re-claim age cases and `test_stale_claim_reclaim_event_records_diagnostic_payload`.

**Non-blocking notes**
- Overwrite semantics mean `started_at` no longer reflects "time since task first became ready" — dashboard "age" will reset on reclaim; if UI needs both "first ready age" and "current worker age", it already has `task_runs`; document which metric dashboard shows to avoid confusion.
- Stale-claim reclaim event now records diagnostic payload — ensure event doesn't log PII from `assignee`/`claim` metadata.

Good fix for elapsed-since-first-claim drift.

Issue #80926, opened 2026-08-07
[Windows/Desktop] In-app update fails with WinError 5 (Permission Denied) due to Hermes-managed Node locking


i0ivi0i:
### Describe the bug
On Windows, attempting to update Hermes via the Desktop App's 1-click update button consistently fails with a `PermissionError: [WinError 5] Access is denied` pointing to `%HERMES_HOME%\node\npm.cmd`.

### Root Cause
This is an architectural lock issue introduced by the new **Hermes-managed Node** feature.
1. The Hermes desktop app relies on the managed Node.js environment located at `AppData\Local\hermes\node` to run.
2. When the user clicks update from within the GUI, the updater script attempts to verify, rebuild, or upgrade this managed Node environment.
3. Because the Electron/Node processes are actively executing from this directory, Windows enforces strict file locks. Any attempt by the update script to touch `npm.cmd` or overwrite the Node tree is immediately rejected by the OS with `WinError 5`.

### Impact
This completely breaks the seamless 1-click update experience for Windows users. They are forced to:
1. Completely shut down the desktop app.
2. Manually kill any lingering background `node.exe` processes via Task Manager / `taskkill`.
3. Open an external terminal (e.g., PowerShell) and run `hermes update`.

### Expected behavior
The updater needs a Windows-safe mechanism for self-updating the active Node runtime. Possible solutions:
- Use a detached bootstrap/updater process that guarantees the desktop app and all Node subprocesses are completely terminated *before* touching the `%HERMES_HOME%\node` directory.
- Download the new Node tree to a side-by-side temporary directory and perform an atomic swap only upon the next cold start.

### Environment
- OS: Windows 10 / 11
- Surface: Hermes Desktop App
- Version: `0.19.0` / `0.20.0`

DavidMetcalfe:
Thanks for the detailed report — the diagnosis is right, and the destructive write code paths are confirmed on current main.

**Root cause (confirmed):** the Hermes-managed Node feature writes destructively into the live tree at `%HERMES_HOME%\node` while the app's Node processes are executing from it:

- The Node 26 requirement (#713a983e4, 2026-08-01) marks every pre-existing tree as "outdated", so the next update/npm resolution triggers the heal — `_heal_managed_node_windows()` did `shutil.rmtree(%HERMES_HOME%\node)` + move.
- The EBADENGINE repair path (`upgrade_managed_npm`) runs `npm install --global --prefix %HERMES_HOME%\node` in place.
[...]

PR #83589, opened 2026-08-11
fix(node): keep an unreadable managed Node tree off PATH

nezuchills:
## What does this PR do?

Stops Hermes from prepending a Node runtime directory it **cannot read** to `PATH`.

`with_hermes_node_path()` and `_managed_runtime_path_entries()` both selected managed runtime dirs with `is_dir()`. An unreadable directory — one whose ACL denies traversal, which an interrupted or elevated install can leave behind at `$HERMES_HOME/node` — still stats as a directory, so it went onto `PATH` anyway.

Windows process creation then **fails on that entry instead of skipping it**. `spawn` returns `EPERM` (errno -4048) as soon as an unreadable directory precedes the real one, so one bad entry breaks *every* child process, not just the ones that wanted a managed runtime.

Reproduced directly on Windows 11 (Node v24.16.0):

```
clean PATH:                        out: v24.16.0
unreadable dir prepended to PATH:  err: EPERM
```

Where it actually bit: the packaged desktop build, with `cross-env` unable to spawn `node` at all.

```
Error: spawn EPERM
    at ChildProcess.spawn (node:internal/child_process:441:11)
    at spawn (node_modules/cross-spawn/index.js:12:24)
```

The fix adds `path_entry_usable()`, which probes by **listing** instead of stat-ing — the only way to distinguish "is a directory" from "is a directory I can traverse". `scandir` raises on the first entry when the dir is unreadable, so it stays O(1).

### Scope — deliberately the PATH half only

The *resolution* half of this problem (bare `Path.is_file()` probes propagating `OSError` out of `find_hermes_node_executable` and friends) is already covered by #70725 and #75419, so this PR does not touch those call sites — no competing duplicate.

Those PRs and this one are complementary, and neither subsumes the other: they stop the resolver from crashing on an unreadable tree, while this stops that same tree from being handed to every child process. #75419 notes that "the caller's job is to keep such a tree out of the PATH" — this is that caller-side fix.

## Related Issue

No existing issue found for the PATH/`EPERM` behaviour (searched issues and PRs for `spawn EPERM`, `with_hermes_node_path`, `_managed_runtime_path_entries`).

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `hermes_constants.py` — add `path_entry_usable()`; use it in `with_hermes_node_path()` instead of `is_dir()`.
[...]

PR #75419, opened 2026-07-31
fix(node): don't let an unstattable path abort Node/npm resolution

twchichi:
## Problem

`hermes dashboard` crashes on Windows when a directory the Node probes scan contains a reparse point Windows cannot traverse:

```
File "hermes_cli/main.py", line 10262, in cmd_dashboard
    if not _build_web_ui(PROJECT_ROOT / "web", fatal=True):
...
File "hermes_constants.py", line 478, in find_hermes_node_executable
    if candidate.is_file() and (
OSError: [WinError 1920] The file cannot be accessed by the system:
  'C:\Users\...\.hermes\node\bin\npm'
```

`Path.is_file()` swallows only the winerrors listed in `pathlib._IGNORED_WINERRORS` — 21 (`NOT_READY`), 123 (`INVALID_NAME`), 1921 (`CANT_RESOLVE_FILENAME`). `ERROR_CANT_ACCESS_FILE` (**1920**) is not in that list, so it propagates instead of reporting "not a file".

How I hit it: a POSIX Node tarball had been unpacked into a Windows `HERMES_HOME`, leaving `node/bin/npm` as a Unix symlink (`../lib/node_modules/npm/bin/npm-cli.js`) and `node/bin/node` as an ELF binary. Any dangling junction, WSL/Docker leftover, or cloud-storage placeholder in a scanned directory reproduces it.

## Why this is worse than a crash

The raise fires before *either* recovery path can run:

- the PATH fallback in `find_node_executable` — a perfectly good `node.exe` was on PATH the whole time; and
- `heal_hermes_managed_node`, which on Windows would have redownloaded a working portable Node over the bad tree. That is the exact situation it exists for, and it never gets reached.

`find_node_executable_on_path` carries the same hazard over a wider surface: it walks **every** `PATH` entry, so a single bad reparse point anywhere on `PATH` takes out the fallback too.

## Fix

Route all six probes through `_is_probe_file`, which maps `OSError` to `False`. No behavior change for paths that stat normally.

## Test

`test_unstattable_managed_entry_is_treated_as_absent` makes `Path.is_file()` raise for one managed entry, then asserts the probe reports the tree absent and resolution falls back to `PATH`. It fails on `main` with the original `OSError` and passes with the fix.

Local run (Windows 11, Python 3.11):

```
tests/test_hermes_constants.py -k ManagedNode  ->  4 passed
tests/test_hermes_constants.py (full file)     ->  38 passed, 6 skipped, 4 failed
```

[...]

PR #97212, opened 2026-08-28
fix(windows): skip unreadable Node candidates instead of crashing hermes update (#97196)

### LINKED ISSUE #97196: Windows: unreadable %LOCALAPPDATA%\hermes\node (ACL denial) crashes update — is_file() probe unguarded, self-heal unreachable
### Summary
On Windows, if the Hermes-managed Node tree at `%LOCALAPPDATA%\hermes\node` becomes **unreadable** (an ACL that denies the current user, as opposed to a file *lock*), `hermes update` dies with an uncaught `PermissionError` before the managed-Node self-heal can run.
This looks like #80926 but is a different cause. #80926 was a **file lock** (a running `node.exe` holding `npm.cmd` open), fixed in #86663 by staging the heal to a sibling dir and rename-swapping. That fix makes the **heal** Windows-safe — but the **probe that runs before it** is still unguarded, so an ACL-denied tree never reaches the healing path at all.
### Environment
- Hermes Agent v0.20.6, install method `git`, `main @ 9978706e`
- Windows 11, Python 3.11.15
- Managed Node tree present but ACL-denied to the invoking user (`Get-Acl` → "Attempted to perform an unauthorized operation")
### Traceback
```
PermissionError: [WinError 5] Access is denied: 'C:\...\AppData\Local\hermes\node\npm.cmd'
  hermes_constants.py, in find_hermes_node_executable._first_runnable
    if candidate.is_file() and (
  pathlib.py, in is_file -> stat -> os.stat
  hermes_cli/main.py, in cmd_update
  hermes_cli/update_cmd.py, in _cmd_update_impl
    _repair_node_deps_on_current_checkout
    _update_node_dependencies
```
### Root cause
Two unguarded `Path.is_file()` calls, both on current `main`:
**1. `hermes_constants.py:931`** — in `find_hermes_node_executable._first_runnable()`:
```python
candidate = directory / name
if candidate.is_file() and (
    sys.platform == "win32" or os.access(candidate, os.X_OK)
):
    resolved = str(candidate)
    if node_tool_runnable(resolved):   # <- catches OSError, but only AFTER the probe
        return resolved, broken
```
`node_tool_runnable()` does handle `OSError`, but it is only reached *after* `is_file()` has already raised. The heal a few lines below (`if needs_heal and heal_hermes_managed_node()`) is therefore unreachable for this failure mode.
**2. `hermes_constants.py:974`** — in `find_node_executable_on_path()`:
```python
for directory in os.environ.get("PATH", "").split(os.pathsep):
    candidate = Path(directory) / name
    if candidate.is_file():
        return str(candidate)
```
This one matters more than it first appears: the installer puts `%LOCALAPPDATA%\hermes\node` **on PATH**, so the PATH search walks the same poisoned directory and dies too. That removes the last fallback — all three resolution entry points fail:
```
find_hermes_node_executable  -> PermissionError
find_node_executable         -> PermissionError
find_node_executable_on_path -> PermissionError   # even the PATH fallback
```
### Reproduction
1. On Windows, deny the current user read access to `%LOCALAPPDATA%\hermes\node` (this arises naturally when the tree is created by an elevated installer run — in my case a self-heal triggered by an `EBADENGINE` failure).
2. Run `hermes update`.
3. Update aborts with the traceback above; the desktop app is never rebuilt.
### Expected
A permission error on one candidate should mean *skip this candidate*, not *abort the update*. With the tree unreadable, resolution should fall through to Node on PATH, or trigger `heal_hermes_managed_node()`.
### Suggested fix
Wrap both probes so an unreadable candidate is skipped rather than fatal — consistent with how `node_tool_runnable()` already treats `OSError`:
```python
try:
    is_file = candidate.is_file()
except OSError:      # PermissionError is a subclass
… omitted 8 diff line(s) across 1 additional file(s)/section(s)

sanjaynandanj:
## Summary

Closes #97196.

On Windows, if the managed Node tree at \`%LOCALAPPDATA%\hermes\node\` becomes **unreadable** by the current user (an ACL that denies the user, as opposed to a file lock), \`hermes update\` dies with an uncaught \`PermissionError\` before the managed-Node self-heal can run:

\`\`\`
PermissionError: [WinError 5] Access is denied:
  'C:\...\AppData\Local\hermes\node\npm.cmd'
    hermes_constants.py, in find_hermes_node_executable._first_runnable
      if candidate.is_file() and (
\`\`\`

Distinct from #80926 (a *file lock* held by a running \`node.exe\`, fixed in #86663 by staging the heal to a sibling dir). That fix makes the **heal** Windows-safe, but the **probe** that decides whether to heal is still unguarded, so an ACL-denied tree never reaches the healing path at all.

The installer also puts \`%LOCALAPPDATA%\hermes\node\` on \`PATH\`, so \`find_node_executable_on_path\` walks the same poisoned tree and dies the same way — knocking out the last fallback.

## Fix

Introduce \`_safe_is_file(candidate)\` — returns \`False\` on any \`OSError\` (\`PermissionError\` is a subclass), consistent with how \`node_tool_runnable\` already treats \`OSError\`. Apply it at:

- \`find_hermes_node_executable._first_runnable\` — lets the existing \`needs_heal\` / \`heal_hermes_managed_node\` path do its job.
- \`find_node_executable_on_path\` (both the has-separator fast path and the PATH walk) — restores PATH as a genuine fallback when the managed tree is unreadable.

\`FileNotFoundError\` maps to \`False\` for free (it's an \`OSError\` subclass and also what \`is_file()\` returns natively for a missing path), so behavior for the common \"tree not present\" case is unchanged.

## Test plan

- [ ] On Windows: reproduce per the issue (deny the current user read access to \`%LOCALAPPDATA%\hermes\node\`) and run \`hermes update\`. Verify no \`PermissionError\` traceback; the self-heal path fires and the update proceeds.
- [ ] On Windows with a healthy managed tree: \`hermes update\` still resolves the managed \`npm.cmd\` (no regression).
- [ ] On POSIX: unchanged — \`os.access(candidate, os.X_OK)\` still gates execution and \`is_file()\` never raises for a missing path.
- [ ] Sanity: \`_safe_is_file\` returns \`False\` for non-existent paths, \`False\` when \`is_file()\` raises \`PermissionError\`, and \`True\` for a regular file.

### DIFF
--- a/hermes_constants.py
+++ b/hermes_constants.py
@@ -804,6 +804,25 @@ def _managed_node_tree_outdated(home: Path | None = None) -> bool:
     return False
 
 
+def _safe_is_file(candidate: Path) -> bool:
+    """``Path.is_file()`` that treats an unreadable candidate as "not a match".
+
+    ``Path.is_file()`` calls ``os.stat()``, which raises ``PermissionError``
+    (``WinError 5``) on Windows when the current user is denied ACL access to
+    the candidate — e.g. ``%LOCALAPPDATA%\\hermes\\node`` populated by an
+    elevated install. An unguarded probe there aborts ``hermes update``
+    before the managed-Node self-heal can run and before the PATH fallback
+    is tried. Skip an unreadable candidate instead of raising, consistent
+    with how :func:`node_tool_runnable` already treats ``OSError``.
+    ``FileNotFoundError`` maps to ``False`` for free (it is an ``OSError``
+    subclass and also what ``is_file`` returns natively for a missing path).
+    """
+    try:
+        return candidate.is_file()
+    except OSError:
+        return False
+
+
 def find_hermes_node_executable(command: str) -> str | None:
     """Return a Hermes-managed Node/npm executable path, healing broken trees.
 
@@ -820,7 +839,7 @@ def find_hermes_node_executable(command: str) -> str | None:
         for directory in iter_hermes_node_dirs():
             for name in names:
                 candidate = directory / name
-                if candidate.is_file() and (
+                if _safe_is_file(candidate) and (
                     sys.platform == "win32" or os.access(candidate, os.X_OK)
                 ):
                     resolved = str(candidate)
@@ -856,14 +875,18 @@ def find_node_executable_on_path(command: str) -> str | None:
         sep and sep in command_str for sep in (os.sep, os.altsep, "/", "\\")
     )
     if has_path_separator:
-        return command_str if Path(command_str).is_file() else None
+        return command_str if _safe_is_file(Path(command_str)) else None
 
     for name in _candidate_node_command_names(command_str):
         for directory in os.environ.get("PATH", "").split(os.pathsep):
             if not directory:
                 continue
             candidate = Path(directory) / name
-            if candidate.is_file():
+            # The installer puts %LOCALAPPDATA%\hermes\node on PATH, so this
+            # walk hits the same tree as find_hermes_node_executable(). An
+            # unguarded is_file() there would let an ACL-denied managed tree
+            # take out the PATH fallback too, leaving no resolution path.
+            if _safe_is_file(candidate):
                 return str(candidate)
     return None
 


Enough1122:

**Overall:** Prevents `hermes` startup crash when `%LOCALAPPDATA%\hermes\node` contains unreadable candidates (Windows elevated installer).

**What it does**
- `hermes_constants.py` adds `_safe_is_file(candidate: Path)` wrapping `Path.is_file()` / `os.stat()` with `PermissionError`/`OSError` handling, treating unreadable candidate as "not a match" instead of raising `WinError 5`. Used by managed-node tree outdated check.
- Mirrors sibling fix in `profiles.py` (#97176) for unreadable skill trees.

**Non-blocking notes**
- Swallowing `PermissionError` is correct for startup resilience, but may hide a real ACL misconfiguration where user *should* fix permissions; consider debug log when skipping unreadable candidate to aid `hermes doctor` forensics (currently silent, like 97176).
- Helper is narrow (`is_file` only) — if other `Path` ops in same flow also `stat` (e.g. `is_dir`, `exists`), they may still raise; audit surrounding `_managed_node_tree_outdated` for similar unguarded calls.

Good Windows robustness fix; no behavior change when readable.

andrexibiza:
The unreadable managed-candidate case is reproduced on current main and coordinated in #104294, alongside ChiChi's earlier #75419 helper, nezuchills' #83589 PATH changes, and Tan Tu's #104212 ACL report. This PR's source SHA `4228188485a9dd2d6e211923d50e8f18d9abd5be` is explicitly credited. The integration retains the present-but-broken state so repair is reached, and the native destination-ACL case now passes. Thanks for helping isolate this boundary. The combined carrier's full repro, tests and merge coordination are in its body.

Issue #73042, opened 2026-07-28
context-usage statusbar item toggle broken when hidden by default (deadlock)


lakitu12:
## Summary

PR #72442 added `context-usage` to `STATUSBAR_HIDDEN_BY_DEFAULT` (commit 5b9518db4). This causes a deadlock: the item is hidden by default and cannot be recovered via the right-click toggle menu.

## Root Cause

The `context-usage` statusbar item has a **runtime** `hidden` condition:

```tsx
// apps/desktop/src/app/shell/hooks/use-statusbar-items.tsx:501
hidden: !contextUsage,
```

Where `contextUsage` comes from `usageContextLabel(currentUsage)`:

```tsx
// apps/desktop/src/lib/statusbar.tsx:43
export function usageContextLabel(usage: UsageStats): string {
  if (usage.context_max) {
    return `${compactNumber(usage.context_used ?? 0)}/${compactNumber(usage.context_max)}`
  }
  return usage.total > 0 ? `${compactNumber(usage.total)} tok` : 
}
```

`$currentUsage` is initialized without `context_max`/`context_used`/`context_percent`:

```tsx
// apps/desktop/src/store/session.ts:361
export const $currentUsage = atom<UsageStats>({
  calls: 0,
  input: 0,
  output: 0,
  total: 0
})
```

The only code path that populates `context_max`/`context_used` is `publishContextUsage` → `ContextUsagePanel.onUsageSnapshot`, which fires when **ContextUsagePanel fetches data from the backend**. But ContextUsagePanel only mounts when the user clicks the statusbar item — which is hidden because `contextUsage` is empty.

**Deadlock:** Item hidden → user toggles it on → runtime `hidden: !contextUsage` still true → item stays invisible → panel never mounts → context data never populated → item stays invisible forever.

## Items NOT affected

`running-timer` (`hidden: !busy || !turnStartedAt`) and `session-timer` (`hidden: !sessionStartedAt`) are **not** affected because their runtime data (`$busy`, `$turnStartedAt`, `$sessionStartedAt`) comes from independent backend events, not from clicking the item itself.

## Suggested fix

Remove `context-usage` from `STATUSBAR_HIDDEN_BY_DEFAULT` in `apps/desktop/src/store/statusbar-prefs.ts`, or alternatively populate `$currentUsage` with context data from an independent source (the periodic status poll) so the item can show once toggled on.

lakitu12:
## Correction

[...]

PR #79175, opened 2026-08-05
fix(desktop): seed context_max so the context-usage statusbar item is visible

Sahil-SS9:
Fixes #78936

## Problem

The context-usage statusbar item is permanently hidden after a session restart. The root cause is a Catch-22:

- `usageContextLabel()` returns empty (so the item's `hidden` is true) whenever `context_max` is unset.
- `context_max` was only populated by `ContextUsagePanel`'s `onUsageSnapshot`, which fires only when the panel mounts — i.e. only when the user clicks the item to open the panel.
- The item is hidden, so it can never be clicked, so `context_max` is never populated.

Normal `message.complete` events only carry `input_tokens`/`output_tokens`, never `context_max`.

## Fix

Eagerly fetch `session.context_breakdown` for the active session and seed `context_max`/`context_percent`/`context_used` into `$currentUsage` once, so the item is visible from the start.

- New `useContextUsageSeed` hook (pure, DI-testable): fetches the breakdown for the active session when `context_max` is unknown, publishes the snapshot via the existing `publishContextUsage` path, and stops once `context_max` is known. Fails open — a failed fetch leaves the item hidden rather than showing bad data. Cancels in-flight fetches on session switch.
- Wired into `useStatusbarItems` after `publishContextUsage` is defined.
- The `ContextUsagePanel` still refetches on open, so the panel and the seeded statusbar stay consistent.

## Verification

- New `use-context-usage-seed.test.ts` (5 tests): seeds once for the active session, skips when `context_max` is known, skips with no active session, fails open on fetch error, and discards stale results after a mid-fetch session switch.
- `npx tsc --noEmit` clean.
- Existing `context-usage-panel.test.tsx` and `statusbar-visibility.test.tsx` still pass.

Issue #78936, opened 2026-08-04
Context usage indicator permanently hidden due to context_max initialization Catch-22


Erich444:
### Bug Description

The context usage indicator icon in the status bar is permanently hidden after a session restart. The root cause is a Catch-22: `context_max` is required for the icon to be visible, but `context_max` is only populated when the user clicks the icon to open the context panel — which is impossible when the icon is hidden.

Normal `message.complete` events only contain `input_tokens`/`output_tokens` — never `context_max`.

### Steps to Reproduce

Start Hermes Desktop
Begin a new session or restart the app
Observe the status bar — the context usage icon is missing
The icon never appears during the session because context_max is never initialized
Expected Behavior: The icon should be visible showing current token usage (e.g. "12.8k tok").

Actual Behavior: The icon is hidden because $currentUsage starts without context_max, so the usageContextLabel() returns empty, and the status bar item remains hidden.

### Expected Behavior

The context usage icon should be visible in the status bar showing current token usage.

### Actual Behavior

The icon is permanently hidden because the app starts without knowing the maximum context size, so it hides the icon to prevent showing incorrect data.

### Affected Component

Other

### Debug Report

```shell
Hermes Desktop: v0.20.0
OS: Windows 11
Backend: Remote llama.cpp server (Qwen3.6-27B-MTP-GGUF)
Diagnosis: Backend API returns usage data correctly; frontend context_max initialization is missing.
```

### Operating System

Windows 11

### Python Version

3.11.15

### Hermes Version

v0.20.0

### Additional Logs / Traceback (optional)

```shell

```

### Are you willing to submit a PR for this?

- [ ] I'd like to fix this myself and submit a PR

OutThisLife:
Verified on main (d350422b): the context gauge no longer self-hides — 25d1c8d740 fetches the breakdown as soon as the gauge is on screen and the item shows a placeholder `—` instead of hiding while values load (use-statusbar-items.tsx:663-666); the never-self-hide wiring is 43e67d872f. Your 2026-08-06 build predates the fix; the gauge remains opt-in by design. Closing as fixed on main.

PR #97214, opened 2026-08-28
fix(desktop): retry context-breakdown fetch until gateway hydration completes

### LINKED ISSUE #97163: [Bug]: Context usage statusbar meter stays hidden after restart until toggled — one-shot breakdown fetch races session hydration (no retry)
### Bug Description

After restarting the Hermes desktop app, the "Context usage" statusbar item (`context-usage` — the `540K/1M [██████░░░░] 54%` meter) does not render, even though it is still checked in the statusbar's right-click menu. Unchecking and re-checking the item makes it appear immediately. Sending a message also recovers it. No error is shown anywhere.

### Steps to Reproduce

1. Enable the context meter: right-click the statusbar → check "Context meter" (上下文用量 in zh).
2. Fully quit the desktop app and relaunch it (cold backend start, ~5 s).
3. Observe: no context meter in the statusbar, although the right-click menu still shows "Context meter" as checked.
4. Uncheck and re-check "Context meter" → the meter appears immediately and stays until the next restart.

### Expected Behavior

The meter renders right after restart, like any other checked statusbar item, without requiring a manual toggle or a new message.

### Actual Behavior

The item is self-hidden for the whole session. In `use-statusbar-items.tsx` the item declares `hidden: !contextUsage` (L549), and `usageContextLabel()` returns `''` whenever `context_max` is falsy and `total === 0` (`apps/desktop/src/lib/statusbar.tsx:44-50`). After a restart both are falsy, because the resumed session is "cold" in the new gateway process: no turn has run in this process, so the streamed usage carries no `context_max`/`total`, and the estimate that should fill the gap never lands (see Root Cause).

### Affected Component

Other

_No response_

### Debug Report

```shell
Report       https://paste.rs/iRcSH
agent.log    https://paste.rs/ZKPfl
gui.log      https://paste.rs/xUH9W
desktop.log  https://paste.rs/CNbau
```

### Operating System

Windows 11

### Python Version

3.11.16

### Hermes Version

backend hermes-agent 0.20.6 (pyproject.toml), desktop app 0.17.0 (apps/desktop/package.json); installed 2026-08-28 from main (ZIP install — git repo has no commits)

```shell

```

The UI covers the cold-session gap with an eager estimate: `useContextBreakdown` (`apps/desktop/src/app/shell/hooks/use-context-breakdown.ts`) fires the `session.context_breakdown` RPC **exactly once** per `[busy, enabled, requestGateway, sessionId]` change (L57). That single fetch races session hydration and can lose in two ways:

1. **RPC lands before the session is in the runtime registry** → `_sess_nowait` returns error `4001 "session not found"` (`tui_gateway/server.py:3374-3394`). The client swallows it: `.catch(() => undefined)` (L47).
2. **RPC lands while the session exists but the agent is not yet hydrated** → the handler returns a *successful* payload with `context_max: 0` (`tui_gateway/methods_session.py:1707-1720`, the `agent is None` branch) — a "successful but empty" response, so no error path fires at all.

In both cases `context_max` stays falsy → the label is `''` → the item hides itself. While the app is idle, none of the effect deps change, so there is no retry and the meter stays hidden for the entire session.

Why the workarounds work: toggling the item flips `enabled` → the effect re-runs → by then the agent is hydrated and the real `context_max` (from `agent.context_compressor.context_length`) comes back. Sending a message has the same effect via the `busy` dep flipping after the turn.

This is the same visible symptom as #78936, but that Catch-22 (context_max only populated by opening the panel) was already addressed by the eager fetch this hook performs; the remaining failure is the one-shot, no-retry race described above.

### Proposed Fix (optional)

In `useContextBreakdown`, treat "no usable data yet" as a transient condition and retry until it resolves:

- Retry on RPC failure **and** on success-with-`context_max === 0`, with bounded backoff (a few attempts a couple of seconds apart), and/or
- Gate the fetch on gateway readiness (the renderer already observes gateway socket state), and/or
- Re-fetch when the session's agent becomes available (e.g., first `busy` transition).

Jackal991:
Closes #97163

## Summary

After a desktop-app restart, the "Context usage" statusbar meter stays hidden until the user toggles the item or sends a message. The eager `session.context_breakdown` fetch in `useContextBreakdown` is one-shot per effect run and loses two different races against gateway hydration:

1. **RPC before the session is registered** → gateway answers error `4001 "session not found"` (`tui_gateway/server.py` `_sess_nowait`). The client swallowed it via `.catch(() => undefined)`.
2. **RPC while the session exists but the agent is not yet hydrated** → the handler's `agent is None` branch (`tui_gateway/methods_session.py:1707-1720`) returns a *successful* payload with `context_max: 0`. The client accepted it and stopped.

In both cases `context_max` stays falsy → `usageContextLabel()` returns `''` → the item hides itself. While the app sits idle, no effect dependency ever changes, so there is no retry and the meter stays hidden for the whole session. Toggling the item (`enabled` flip) or sending a message (`busy` flip) re-runs the effect — which is exactly the reported workaround behavior.

## Changes

In `apps/desktop/src/app/shell/hooks/use-context-breakdown.ts` (client-side only):

- Treat "no usable data yet" as transient: `isUsable()` requires `context_max > 0` (a zero ceiling means the backend has no hydration data yet, not an empty context — a session always has at least a system prompt below it).
- Bounded retry on **both** failure modes: rejection and success-with-unusable-data. `RETRY_LIMIT = 4` retries at `RETRY_DELAY_MS = 1500` (max 5 RPCs over ~6s, then gives up — no infinite polling).
- Retry budget is per effect run: the attempt counter lives in the effect closure, so a dependency change (session switch, toggle, turn boundary) grants a fresh budget, matching the hook's existing re-fetch semantics.
- Discipline preserved: `cancelled` checked after every await and before every setState; `loading` start/finish via `.finally`; return shape and keyed-by-sessionId semantics unchanged; public signature identical.

## Tests

New `use-context-breakdown.test.tsx` (5 tests, fake timers, ~1.4s):

- **4001 recovery**: rejection once, then success → hook ends with `context_max` set. Fails on the old code.
- **empty-success recovery**: first success with `context_max: 0`, then real data → hook retries instead of painting an empty gauge. Fails on the old code.
- **bounded**: always rejects → exactly 5 calls, then silence (verified over a 60s timer advance), `loading` false.
- **no cross-session leak**: deferred session-A answer landing after a switch to B never paints under B.
- **fresh budget after dep change**: `busy` flip mid-retry cancels the old timer (cleanup) and the new run accepts its first success.

Revert-to-buggy proof: with the fix stashed and only the test file present, 4/5 tests fail (the fifth asserts the unchanged keyed-session behavior). Restored, all 5 pass.

- `npx vitest run --project ui src/app/shell/hooks/use-context-breakdown.test.tsx` → 5/5 pass
- `npx tsc --noEmit` → clean
- `npx eslint` on both touched files → clean; prettier --check clean
- Sibling suites still green: `context-usage-panel.test.tsx` + `use-status-snapshot.test.ts` (14 passed), `statusbar-visibility.test.tsx` + `statusbar-context-menu.test.tsx` (12 passed)

## Coordination with related open PRs

- **#79175** (targets the older #78936 Catch-22) does **not** fix this race: it adds a separate `useContextUsageSeed` hook that fails open on fetch errors and publishes `context_max` unconditionally (including `0`), and it does not touch `use-context-breakdown.ts`. This PR is complementary — it fixes the fetch inside `useContextBreakdown` itself.
- Three open PRs (#79175, #87925, #89839) touch the same hook area, so merge conflicts are possible; happy to rebase.
- The issue also suggested gateway-readiness gating and refetch-on-busy-transition as alternatives. The bounded retry was chosen because it covers both failure modes client-side with no backend changes and no new state coupling; the existing `busy` dep already provides the refetch-on-turn boundary.

### DIFF
--- a/apps/desktop/src/app/shell/hooks/use-context-breakdown.test.tsx
+++ b/apps/desktop/src/app/shell/hooks/use-context-breakdown.test.tsx
@@ -1,0 +1,174 @@
+import { act, cleanup, renderHook } from '@testing-library/react'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import type { ContextBreakdown } from '@/types/hermes'
+
+import { deferred } from '../../../test/deferred'
+
+import { useContextBreakdown } from './use-context-breakdown'
+
+type GatewayRequester = <T = unknown>(method: string, params?: Record<string, unknown>) => Promise<T>
+
+function breakdownOf(contextMax: number): ContextBreakdown {
+  return {
+    categories: [{ color: 'teal', id: 'conversation', label: 'Conversation', tokens: 24_100 }],
+    context_max: contextMax,
+    context_percent: 9,
+    context_used: 24_100,
+    estimated_total: 25_400,
+    model: 'test-model'
+  }
+}
+
+async function flushAsync() {
+  await act(async () => {
+    await vi.advanceTimersByTimeAsync(0)
+  })
+}
+
+beforeEach(() => {
+  vi.useFakeTimers()
+})
+
+afterEach(() => {
+  cleanup()
+  vi.useRealTimers()
+  vi.restoreAllMocks()
+})
+
+describe('useContextBreakdown retry', () => {
+  it('recovers from a session-not-found rejection once the gateway registers the session', async () => {
+    // The gateway 4001s until its runtime registry knows the session — the
+    // restart race. One failed fetch, then a real answer.
+    const requestGateway = vi
+      .fn()
+      .mockRejectedValueOnce(new Error('session not found'))
+      .mockResolvedValue(breakdownOf(123_000))
+
+    const { result } = renderHook(() =>
+      useContextBreakdown({ busy: false, enabled: true, requestGateway, sessionId: 'runtime-1' })
+    )
+
+    await flushAsync()
+
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(5_000)
+    })
+
+    expect(requestGateway).toHaveBeenCalledTimes(2)
+    expect(result.current.breakdown?.context_max).toBe(123_000)
+    expect(result.current.loading).toBe(false)
+  })
+
+  it('refetches when the first answer has no hydration data instead of painting an empty gauge', async () => {
+    // The session is registered but the agent is not hydrated yet: the RPC
+    // SUCCEEDS with context_max 0. Accepting that answer would strand the
+    // gauge at zero until something else changes.
+    const requestGateway = vi.fn().mockResolvedValueOnce(breakdownOf(0)).mockResolvedValue(breakdownOf(999_000))
+
+    const { result } = renderHook(() =>
+      useContextBreakdown({ busy: false, enabled: true, requestGateway, sessionId: 'runtime-1' })
+    )
+
+    await flushAsync()
+    expect(result.current.breakdown).toBeNull()
+
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(5_000)
+    })
+
+    expect(requestGateway).toHaveBeenCalledTimes(2)
+    expect(result.current.breakdown?.context_max).toBe(999_000)
+    expect(result.current.loading).toBe(false)
+  })
+
+  it('stops after a bounded number of attempts instead of retrying forever', async () => {
+    const requestGateway = vi.fn(async (_method: string) => {
+      throw new Error('session not found')
+    })
+
+    const { result } = renderHook(() =>
+      useContextBreakdown({ busy: false, enabled: true, requestGateway, sessionId: 'runtime-1' })
+    )
+
+    await flushAsync()
+    expect(requestGateway).toHaveBeenCalledTimes(1)
+
+    await act(async () => {
+      await vi.advanceTimersByTimeAsync(60_000)
+    })
+
+    // RETRY_LIMIT is 4 retries: 1 initial fetch + 4 retries — then the hook
+    // gives up and never schedules another attempt, however long it waits.
+    expect(requestGateway).toHaveBeenCalledTimes(5)
+    expect(result.current.breakdown).toBeNull()
+    expect(result.current.loading).toBe(false)
+  })
+
+  it('does not report session A numbers once the focused session is B', async () => {
+    const sessionA = deferred<ContextBreakdown>()
+
+    const requestGateway = vi.fn((method: string, params?: Record<string, unknown>) => {
+      if (params?.session_id === 'runtime-2') {
+        return new Promise<ContextBreakdown>(() => undefined)
+      }
+
+      return sessionA.promise
+    }) as unknown as GatewayRequester
+
+    const { rerender, result } = renderHook(
+      ({ sessionId }) => useContextBreakdown({ busy: false, enabled: true, requestGateway, sessionId }),
+      { initialProps: { sessionId: 'runtime-1' } }
+    )
+
+    rerender({ sessionId: 'runtime-2' })
+    await flushAsync()
+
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Retries context-breakdown fetch until gateway hydration completes — fixes blank gauge on early session load.

**What it does**
- `apps/desktop/src/app/shell/hooks/use-context-breakdown.ts` retries a few times with short gap when RPC errors or returns `context_max: 0` (session registered but agent not yet hydrated), racing gateway startup. Both are not real answers, so retry before giving up.
- `apps/desktop/src/app/shell/hooks/use-context-breakdown.test.tsx` (174 lines) pins retry on `context_max: 0` and on RPC error, using `deferred` + `renderHook`.

**Non-blocking notes**
- Retry is bounded ("a few times with short gap") — verify constants (e.g. 3 retries × 200ms) are exported or documented so tuning doesn't require reading hook source; and that retries are cancelled on unmount to avoid `setState` on unmounted component (test checks cleanup).
- `context_max: 0` could also be a legitimate value for a zero-context model; treating it as "not hydrated" may cause unnecessary retries for such models. If zero-context models exist, consider distinguishing via separate `hydrated: boolean` flag.

Fixes early-lifecycle blank gauge without changing steady-state.

Jackal991:
Thanks. The overall read is right — retry until hydration completes, bounded, no steady-state change.

On the two non-blocking notes:

**1. Constants + unmount cancellation.** Both are already handled. The constants are named and documented at the top of the hook: `RETRY_LIMIT = 4` and `RETRY_DELAY_MS = 1_500` (not "3 × 200ms" as the review guessed), with a comment explaining the budget. Unmount cancellation is covered by the `cancelled` flag checked after every `await` (including after `wait(RETRY_DELAY_MS)`), and the `stops after a bounded number of attempts` test plus the `does not report session A numbers once the focused session is B` test pin the cleanup behavior. So no `setState` on an unmounted component, and tuning doesn't require reading the hook body — the constants are right there.

**2. `context_max: 0` as a legitimate value.** This is the one point I'd push back on. A session always has at least the system prompt below it, so a zero ceiling is not a real answer for any model — it means the backend hasn't hydrated yet. The `isUsable()` helper documents exactly that: "A zero ceiling means the backend has no hydration data yet, not an empty context — a session always has at least the system prompt below it." If a zero-context model ever existed, the retry would just burn a few extra 1.5s gaps before settling on the same `0` — harmless, and it self-corrects. I don't think a separate `hydrated` flag earns its keep for a case that doesn't exist today.

gerdosi:
## Local confirmation (macOS Desktop, still present after #86507)

This is still the live behaviour on a current git install.

**Environment**
- Hermes Agent v0.21.0 (2026.8.31), upstream `5a8e8a6b`
- Hermes Desktop on macOS 27.0 (build 26A5425a), Apple Silicon
- Context meter enabled in **Show in status bar** (checkbox stays ticked)

**What we see**
Switching between **existing** chats that already have real occupancy still blanks the meter. Example: coming into a long thread, the status bar had **no** context chip. Untick → tick **Context meter** (same menu) immediately restored **~270k tokens / ~54%** — not zeros, not an empty session.

Confirmed across several non-empty threads. The preference is remembered; only the chip is missing until that toggle.

**Why this matches the PR**
`hidden: !contextUsage` still unmounts the item when `context_max` is falsy. Session switch drops the previous breakdown on purpose; the one-shot `session.context_breakdown` fetch then loses the hydration race (or a swallowed error). Nothing retries until `enabled` flips. That is exactly the untick/tick workaround.

Related: closed #73042 / merged #86507 fixed “toggle does nothing until you send a message,” but not this **session-switch + already-measured occupancy** case.

Happy to retest a build of this PR.

PR #97220, opened 2026-08-28
fix(updater): stop Scheduled Task gateways without stopping Task Scheduler

### LINKED ISSUE #97208: [Bug] Windows updater tries to stop Task Scheduler ('Schedule' service) when gateway runs via Scheduled Task — Access denied, update aborts
### Symptom

Desktop Update button fails with `Hermes update did not finish. Update failed (exit 1)`. `desktop-update-handoff.log` shows:

```
→ Fleet: 1 running service(s) across profiles: derek
→ Stopping Windows gateway process(es) before updating Hermes...
RuntimeError: Could not stop Windows gateway service Schedule: [SC] OpenService FAILED 5:
Access is denied.
  ...
  File "hermes_cli\update_cmd.py", line 6023, in _pause_windows_gateways_for_update
    _stop_windows_gateway_service(
  File "hermes_cli\update_cmd.py", line 5689, in _stop_windows_gateway_service
    raise RuntimeError(detail or f"sc.exe stop failed with {result.returncode}")
RuntimeError: [SC] OpenService FAILED 5:
```

The updater attempts `sc.exe stop Schedule` — the **Windows Task Scheduler service** — as if it were a Hermes-owned gateway service.

### Repro (100% deterministic)

1. Windows 11, git install at `C:\Users\<user>\AppData\Local\hermes\hermes-agent`
2. Gateway installed as an autostart **Scheduled Task** (`schtasks /query` shows `\HermesGateway` → `cmd.exe /c hermes-gateway-boot.cmd`, running every minute as the user)
3. Gateway process tree: `python.exe -m hermes_cli.main --profile <p> gateway run --replace`, parent chain terminates at `svchost.exe` hosting the `Schedule` service
4. Click Update in the Desktop app (or run `hermes update`)

Result: `_pause_windows_gateways_for_update` → `find_windows_gateway_services` walks the gateway PID's parent chain, finds PID = Task Scheduler's service PID in the SCM table, classifies the gateway as "service-supervised", and `_stop_windows_gateway_service("Schedule")` fails with OpenService access denied → whole update aborts. Happens on **every** attempt while the task-launched gateway is alive; both the Desktop handoff and the CLI retry hit it.

Related to #96860 (same `_pause_windows_gateways_for_update` preflight; that one dies earlier in `find_windows_gateway_services` on unrelated services in transitional states — BITS `start_pending` here). Both stem from treating "parent chain reaches an SCM service" as "Hermes owns that service."

### Root cause

`find_windows_gateway_services` matches any running SCM service whose PID appears in a gateway process's ancestor chain. A gateway launched by a Scheduled Task has `svchost.exe` (Schedule) as its ultimate ancestor, so Task Scheduler is misidentified as the supervising Hermes service.

### Suggested fix

Require positive Hermes ownership before attempting `sc.exe stop`, e.g.:
- Match the service's binary path / `PathName` against the Hermes install (venv `hermes.exe`, `hermes-gateway-boot.cmd`, or the hermes-agent checkout), and/or
- Only consider services created by Hermes's own service-installer naming (`hermes-gateway*`), and/or
- Exempt the well-known Windows service set (`Schedule`, `BITS`, `Winmgmt`, ...) from ever being classified as gateway supervisors.

A task-launched gateway should be paused by killing its PID (which the updater already knows) and resumed via the task's own next trigger, not by stopping the OS's scheduler.

### Workaround

Disable the Scheduled Task (elevated: `schtasks /change /tn HermesGateway /disable` + `schtasks /end /tn HermesGateway`), run the update, re-enable. Verified working on this machine.

### Environment

- Hermes Agent at `a641644f1` (main) · Windows 11 · Python 3.11.15
- HERMES_HOME `C:\Users\kaize\AppData\Local\hermes`
- Gateway profile `derek` via Scheduled Task `HermesGateway`

fangliquanflq:
## What does this PR do?

Windows updates no longer try to stop the operating system's Task Scheduler service when a Hermes gateway was launched by a Scheduled Task. Service-managed gateways still use SCM pause and resume when the service explicitly declares Hermes ownership in its name.

### Symptom

With a gateway launched by the `HermesGateway` Scheduled Task, `hermes update` can abort after trying `sc.exe stop Schedule`, which fails with `OpenService FAILED 5: Access is denied`.

### Impact

Affected Windows users cannot complete Desktop or CLI updates while the task-launched gateway remains alive.

### Bug Cause

**Trigger:** `hermes_cli/gateway.py:find_windows_gateway_services()` selects the first unique SCM service PID in a validated gateway process's ancestor chain.

**Causal chain:**
1. Windows Task Scheduler launches the gateway process from a Scheduled Task.
2. The gateway ancestor chain reaches the `svchost.exe` process hosting the `Schedule` service.
3. SCM discovery treats that ancestry alone as service ownership and the updater calls `sc.exe stop Schedule`, aborting on access denied.

**Why it is wrong:** Process ancestry proves that a service participated in launching a process, but it does not prove that the service is a Hermes-owned supervisor that the updater may stop.

**Working sibling / contrast:** A genuinely service-managed gateway named `HermesGateway` carries an explicit Hermes ownership signal and remains eligible for SCM pause and resume.

**Ruled out:** PID reuse is not the cause. The existing gateway and service creation-time checks pass in this scenario because both PIDs are live and stable; the missing condition is positive service ownership.

### Fix

Normalize SCM service names and consider only `Hermes` or `HermesGateway*` names as Hermes-owned supervisors. Add a regression test whose validated gateway PID is a descendant of the unique `Schedule` service PID and verify that SCM discovery returns no owned service.

## Related Issue

Closes #97208

## Type of Change

## Changes Made

- `hermes_cli/gateway.py` - require an explicit Hermes service-name ownership signal before mapping an SCM ancestor to a gateway.
- `tests/hermes_cli/test_gateway.py` - cover a task-launched gateway under the Windows `Schedule` service while preserving real and ambiguous Hermes service cases.

## How to Test

1. Launch a Hermes gateway through the `HermesGateway` Scheduled Task on Windows.
2. Run `hermes update` and verify the updater pauses the gateway PID without invoking `sc.exe stop Schedule`.
3. Run the automated regression suites:

```bash
scripts/run_tests.sh tests/hermes_cli/test_gateway.py -q
scripts/run_tests.sh tests/hermes_cli/test_update_concurrent_quarantine.py -q
```

Local result: 62 passed, 5 skipped.

### Code

- [x] I've run the repository test wrapper on the relevant suites and all tests pass
- [x] I've tested the automated regression on Windows 11

- [x] Documentation update is not applicable
- [x] Config example update is not applicable
- [x] Contributor or agent guide update is not applicable
- [x] I've considered cross-platform impact; the ownership filter is Windows-only
- [x] Tool description or schema update is not applicable

## Screenshots / Logs

N/A


Enough1122:

**Overall:** Allowlist fix: Windows gateway detection only kills Hermes-owned SCM services (#97208 sibling to #97226).

**What it does**
- `hermes_cli/gateway.py:_is_hermes_windows_gateway_service` normalizes `name.casefold()` stripped of non-alnum and checks `== "hermes"` or `startswith "hermesgateway"`.
- `find_windows_gateway_services` now `continue`s when service name not in allowlist, so `Schedule` (Task Scheduler svchost) ancestor no longer counted as gateway host. Existing `pid<=0` / `indeterminate status` fail-closed paths unchanged.

**Non-blocking notes**
- Allowlist is strictest correct fix — guarantees only `Hermes`/`HermesGateway*` services are stopped. Verify service naming is stable; if future installer uses `Hermes Gateway` with space, `casefold+isalnum` collapse (`hermesgateway`) still matches.
- Collides with #97226's denylist approach (`_EXEMPT_WINDOWS_SYSTEM_SERVICES`); both solve `Schedule` but diverge on policy. Merging both would prefer allowlist (this PR) as more fail-closed; if both land, one will be redundant. Coordinate with reviewer to pick one strategy before merge.
- Test `test_find_windows_gateway_services_ignores_task_scheduler_ancestor` uses `FakeService.as_dict` vs newer `name()/pid()/status()` API — ensure shim still matches current `win_service_iter` contract.

Minimal and correct; ties to #97226 discussion.

TomSpoct:
**Live confirmation of the root cause and this fix approach**, from a real-world hit on a Windows 11 laptop (2026-08-31):

### Reproduction
Gateway runs under a Windows Scheduled Task (`Hermes_Gateway`). The updater's gateway-pause preflight (`find_windows_gateway_services`, hermes_cli/gateway.py) walks the process tree:

```
python.exe (gateway run, uv cpython)
  → python.exe (venv gateway run)
    → powershell.exe (Hermes_Gateway_Supervised.ps1)
      → wscript.exe (Hermes_Gateway_Watchdog.vbs)
        → svchost.exe (hosts the "Schedule" Task Scheduler service)
          → services.exe → wininit.exe
```

The ancestor walk mapped the sole service on the svchost host — `Schedule` — as the gateway's supervising service, then attempted `sc.exe stop Schedule`:

```
RuntimeError: Could not stop Windows gateway service Schedule:
[SC] OpenService FAILED 5:
Access is denied.
```

Update aborted with exit 1 **before any fetch/pull**, and the desktop re-triggered in a loop (3 attempts in one morning; receipt on file). Both #98439 and #97208 describe exactly this.

### Fix verified live
I applied a local equivalent of this PR's allowlist approach (skip any service that doesn't declare Hermes ownership during enumeration), plus tests, and ran the previously-failing update end-to-end:

- Gateway-pause preflight now returns `[]` for the Scheduled-Task tree — ` Schedule` is never treated as an owner, the existing unmapped/force-kill + respawn path handles it instead.
- Full update completed successfully: gateway respawned on updated code, no `.update-incomplete` markers, no re-trigger after desktop relaunch.
- The real supervisor (the Scheduled Task) is left untouched and the watchdog respawns the gateway correctly.

### Additional edge cases worth covering in tests (both handled by this allowlist)
1. **Shared svchost host**: `Schedule` + `HermesGatewayA` on the same PID — after allowlist filtering only the Hermes service remains, so the shared-host ambiguity check doesn't mis-fire.
2. **Indeterminate service status** (`STOP_PENDING`): a transitional/stopping `Schedule` is skipped by name before the fail-closed indeterminate-status check can abort the update.

One note on the existing test shim: `test_find_windows_gateway_services_ignores_task_scheduler_ancestor` uses `FakeService.as_dict()`, while newer tests in the file use the `name()/pid()/status()` accessor API — worth aligning to the accessor pattern so it matches the current `win_service_iter` contract (as the automated review flagged).

Verified against all 3 scenarios: 8/8 `find_windows_gateway_services` tests and 47 update-path tests passed. This approach (allowlist, more fail-closed) is the right one to land IMO. Happy to contribute the edge-case tests above as a follow-up commit if useful.

fangliquanflq:
Thanks for the live verification and the concrete edge cases. I updated the PR on top of current `main` and incorporated the actionable points:

- Non-Hermes service names are now rejected before their status or PID is queried, so an inaccessible or transitional `Schedule` service cannot abort SCM discovery.
- Added coverage for a transitional `Schedule` ancestor and for `Schedule` sharing a host PID with `HermesGatewayA`; only the Hermes-owned service remains eligible.
- Preserved the fail-closed behavior added on `main`: a transitional `HermesGateway` ancestor still aborts discovery, and two Hermes-owned services on one host PID remain ambiguous.
- Kept both supported psutil service shapes. Existing `as_dict()` coverage remains intentional, while the shared-host and inaccessible-detail tests exercise the current `name()` / `status()` / `pid()` accessor API.

The earlier reference to #97226 does not represent a competing denylist implementation: that PR currently concerns delegation worktree anchoring and is unrelated to Windows SCM discovery.

Validation on commit `bf3405c5712f89cdded9f01d6d8aed2bc283201f`:
- `scripts/run_tests.sh tests/hermes_cli/test_gateway.py -q` — 37 passed, 5 skipped
- `scripts/run_tests.sh tests/hermes_cli/test_update_concurrent_quarantine.py -q` — 29 passed
- Independent scan-only review — perfect, no findings

Entmarketingteam:
Independent reproduction of this bug on a second machine, in case it helps prioritize. Hit it today in the wild (not synthetically) on Windows 11 26200, Hermes v0.20.6, non-elevated PowerShell:

```
RuntimeError: Could not stop Windows gateway service Schedule: [SC] OpenService FAILED 5:

Access is denied.
```

raised from `_pause_windows_gateways_for_update()` → `_stop_windows_gateway_service("Schedule")`, aborting the update **before the fetch**, exactly as described here.

### Confirming the causal chain on this host

The gateway is started by a Scheduled Task whose action is a PowerShell script that runs the gateway through a secrets-manager wrapper, so the live ancestry at failure time was:

```
hermes.exe gateway run
  └─ <secrets wrapper>.exe          (stays alive)
      └─ powershell.exe             (the Scheduled Task action — stays alive, because it invoked the
                                     wrapper synchronously rather than spawning it)
          └─ svchost.exe            (host of the Schedule service)
```

The single-service condition in `find_windows_gateway_services()` is satisfied here because the Task Scheduler host is not shared:

```
Schedule svchost pid 3168 hosts running services: ['Schedule']
```

That looks like ordinary svchost service splitting on machines above the split threshold rather than anything unusual about this box, which suggests the "exactly one service" branch is the common case for `Schedule` on modern Windows, not an edge case.

### Direct before/after probe

Calling the discovery function against the live gateway PID, same machine, minutes apart:

* Task-launched PowerShell **alive** → the ancestor walk reaches the `Schedule` svchost and it is selected as the supervising service → `sc.exe stop Schedule` → access denied.
* Task-launched PowerShell **exited** (ancestry then terminates at the wrapper process, whose own parent is dead) → `find_windows_gateway_services()` returns `[]` and `hermes update` completes normally, exit 0.

So ancestry alone is doing all the work in the current selection logic, which matches the "process ancestry proves a service participated in launching, not that it supervises" framing in this PR.

### Impact note

While a task-launched gateway is alive there is no non-elevated path through the updater at all — it fails at the same point every time, before fetching. The only workaround is to stop the gateway and disable the launching task first, which is not discoverable from the error message.

### On the two open PRs

Noting that #99361 targets the same bug with a `service_name.casefold() == "schedule"` skip. From this repro I'd favor the allowlist approach in this PR: the failure mode is "an ancestor svchost that *launched* the process is treated as its supervisor", and `Schedule` is only the most common instance of that class, so a name-specific skip leaves the general case open. Requiring the service name to declare Hermes ownership also lines up with the names the installer actually registers (`hermes-gateway` / `Hermes_Gateway`).

Happy to run a patched build against the real Scheduled-Task setup here if a Windows confirmation on the fix would be useful.

fangliquanflq:
Thanks for the independent reproduction and the before/after ancestry probe. I checked the current head (`bf3405c5712f89cdded9f01d6d8aed2bc283201f`), and this report does not require another code change:

- SCM discovery rejects every non-Hermes service name before reading its status or PID, so `Schedule` cannot be selected or passed to `sc.exe stop`.
- The normalization already accepts the reported `hermes-gateway` and `Hermes_Gateway` naming forms.
- Regression coverage includes a running `Schedule`-only ancestor, a transitional `Schedule`, and a shared `Schedule`/`HermesGatewayA` host while retaining fail-closed behavior for ambiguous or transitional Hermes-owned services.
- #99361 applies a narrower `Schedule`-specific exclusion; this PR addresses the broader ownership class demonstrated by your report.

I reran the relevant suites with the repository wrapper:
- `scripts/run_tests.sh tests/hermes_cli/test_gateway.py -q` — 37 passed, 5 skipped
- `scripts/run_tests.sh tests/hermes_cli/test_update_concurrent_quarantine.py -q` — 29 passed

A previous contributor has also live-verified an equivalent allowlist against a Scheduled-Task gateway end to end, so an additional patched run would be useful corroboration but is not required to address this report. Thanks for offering it.

Finalthief:
Third data point, on an axis not yet covered in this thread: **Windows 11 ARM64** (Snapdragon X, build 26200), Hermes **v0.21.0** at `c56f8cdd48`.

To be clear about what this is and isn't: not a fresh reproduction — the two reports above already cover the `sc.exe stop Schedule` / `OpenService FAILED 5` path better than I could restate it. What I can add is that this install has been carrying an out-of-tree patch for this exact bug, and that current `main` still has no guard. `find_windows_gateway_services` walks `gateway_process.parents()` and maps any running SCM service PID onto the gateway, so a Scheduled-Task-launched gateway still resolves `Schedule` as its owner at that commit.

One environment shape that may be worth keeping in the test matrix — two profiles, only one of them task-launched:

- `default` → Scheduled Task → ancestry reaches the `Schedule` svchost (the affected path)
- second profile → Startup-folder `.vbs` → no SCM ancestor at all

So a single install can have one gateway that hits this and one that never does, which makes it easy to see a partial failure and conclude the updater "mostly works."

On the fix itself: we had independently converged on a narrower denylist (skip service names in `{"schedule"}`). Having now read this PR, the allowlist here is strictly better and I'd rather see this one land than what we had. A denylist only ever covers the broker you already found, whereas requiring positive `Hermes`/`HermesGateway*` ownership also rejects any other service that can legitimately appear in an ancestor chain. Rejecting non-Hermes names *before* reading status or PID also closes the transitional-state path that our version left open — our guard ran after the status check, so an inaccessible or `stop_pending` `Schedule` could still abort discovery.

Happy to drop the local patch once this merges.

teknium1:
Thanks @fangliquanflq — this landed on `main` via #115680 (44945d224c): your commit(s) were cherry-picked with authorship intact. The Windows half was proven red→green on a real windows-latest runner (receipts in #115680's body). Closing this one in favour of the merged PR.

PR #93478, opened 2026-08-24
feat(guardrails): pre-tool-call required-argument validation to break empty-shell retry loops

fred0m:
## Summary

Adds a pre-tool-call required-argument validation layer to `tool_loop_guardrails`. Tool calls missing arguments the tool itself declares mandatory are now blocked **before execution**, with a message that names the missing arguments and carries the schema's own guidance for each — so the block message doubles as a repair hint the model can act on immediately.

Motivation: in a real session while deploying a service (self-hosted setup), the agent called `skill_manage create` **nine times in a row** without `content` (one retry even passed `file_content`, the wrong key). Every call executed, failed, and re-entered loop detection; the built-in `Tool loop warning` only fires at 9 same-tool failures and is a warning, not a halt. Validating required args before dispatch cuts the empty-shell retry cycle at the source — no failed execution, no retry, no loop.

## Changes

- **`model_tools.py`** — new `validate_required_tool_args()`:
  - Universal check: `parameters.required` from `ToolRegistry.get_schema(tool_name)`.
  - Conditional check: a `conditional_required` rule table (config) for tools whose schemas are intentionally loose — e.g. `skill_manage` only declares `["action", "name"]` as required, while `content` is required for `create`/`edit`, `old_string`/`new_string` for `patch`, `file_path`/`file_content` for `write_file`.
  - Message: `Tool X is missing required argument(s): Y. <schema description for Y>`.
- **Placement** — inside `handle_function_call`, after plugin `pre_tool_call` hooks (plugins keep priority to `modify`/supply args) and before ACP edit approval. Blocked calls emit `post_tool_call` with `status="blocked"`, `error_type="missing_required_args"`, so they are **not** counted as failures by loop guardrails.
- **`config_defaults.py` / `agent/tool_guardrails.py`** — `pre_tool_validation` (default `true`) and `conditional_required` (overridable; falls back to defaults when absent) under `tool_loop_guardrails`. No new env vars; no change to the tool schemas sent to the model.

Boundary choice: a missing key or explicit `None` is treated as missing; an empty string is not (e.g. `skill_manage patch new_string=""` legitimately deletes matched text).

## Testing

[...]

PR #97222, opened 2026-08-28
feat(agent): post-execution tool result validation

thewulf7:
## What does this PR do?

Adds a lightweight validator that runs **after** every tool call, before the result reaches the LLM. When a tool returns `None`, the wrong type, or a shape that makes no sense for that tool, the agent currently feeds the garbage directly to the model — which then hallucinates a corrective action instead of recognising the failure cleanly.

This PR surfaces shape mismatches early without ever blocking execution:

- Validation failure → **WARNING log + entry in `middleware_trace`** — the original result is always passed through unchanged.
- Unknown / MCP / future tools are **always allowed** through (no rule = no rejection).
- A bug in the validator itself **cannot crash** tool execution (try/except in `tool_executor.py`).

## Related Issue

Complements #93478 (pre-call required-argument validation): that PR stops the LLM dispatching a malformed *call*; this PR detects when a well-formed call returns a malformed *response*.

## Type of Change

## Changes Made

- `agent/tool_result_validator.py` — new module; pure `validate_tool_result(tool_name, result) -> (bool, str | None)`
- `agent/tool_executor.py` — 22-line patch after `_invoke_tool()` returns, before `_emit_terminal_post_tool_call`
- `tests/agent/test_tool_result_validator.py` — 25 unit tests covering all registered tools, unknown tools, and edge cases

## How to Test

```bash
uv run --extra dev python3 -m pytest tests/agent/test_tool_result_validator.py -v
# → 25 passed
```

To see it fire in a live session: run any tool that returns `None` (e.g. mock `read_file` in a test agent) and watch for the WARNING in the log.

### Code

- [x] I searched for existing PRs — #93478 is the closest; this is post-exec, not pre-exec
- [x] I've run `pytest tests/agent/test_tool_result_validator.py -v` — 25 passed
- [x] Tested on: macOS 15 (Apple Silicon)

- [x] Module docstring documents design decisions — or N/A
- [x] No new config keys added — or N/A
- [x] AGENTS.md unchanged (no architecture change, just a new hook point) — or N/A

### DIFF
--- a/agent/tool_executor.py
+++ b/agent/tool_executor.py
@@ -1216,8 +1216,9 @@ class _ConcurrentBatch:
                     "preview": repr(result)[:200],
                 })
         except Exception as _val_err:
-            # Validation is best-effort — never block the tool result
-            logger.debug("tool result validator raised for %s: %s", ref.name, _val_err)
+            # Validation is best-effort — never block the tool result.
+            # Log at WARNING so a bug in the validator itself is visible.
+            logger.warning("tool result validator raised for %s: %s", ref.name, _val_err)
         is_error, _ = _detect_tool_failure(ref.name, result)
         if is_error:
             logger.info("tool %s failed (%.2fs): %s", ref.name, duration, result[:200])
@@ -1439,6 +1440,7 @@ def execute_tool_calls_concurrent(agent, assistant_message, messages: list, effe
     batch = _ConcurrentBatch(agent, messages, effective_task_id, parsed_calls, timeout_s)
     agent._current_tool = tool_names_str
     agent._touch_activity(f"executing {num_tools} tools concurrently: {tool_names_str}")
+
 
 
     spinner = _start_quiet_tool_spinner(agent, "", {}, label=f"⚡ running {num_tools} tools concurrently")
--- a/agent/tool_executor.py
+++ b/agent/tool_executor.py
@@ -1198,6 +1198,26 @@ class _ConcurrentBatch:
         duration = time.time() - start
         if not blocked and not dispatched:
             ref.emit_post(agent, result, duration_ms=int(duration * 1000))
+        # Post-execution tool result validation — catches malformed results before
+        # they reach the LLM, logs a warning, and records the issue in middleware_trace.
+        # Does not block execution: the original result is always passed through.
+        try:
+            from agent.tool_result_validator import validate_tool_result
+            is_valid, validation_error = validate_tool_result(ref.name, result)
+            if not is_valid:
+                logger.warning(
+                    "tool %s returned unexpected result shape: %s",
+                    ref.name,
+                    validation_error,
+                )
+                ref.trace.append({
+                    "type": "tool_result_validation_error",
+                    "error": validation_error,
+                    "preview": repr(result)[:200],
+                })
+        except Exception as _val_err:
+            # Validation is best-effort — never block the tool result
+            logger.debug("tool result validator raised for %s: %s", ref.name, _val_err)
         is_error, _ = _detect_tool_failure(ref.name, result)
         if is_error:
             logger.info("tool %s failed (%.2fs): %s", ref.name, duration, result[:200])
@@ -1419,6 +1439,7 @@ def execute_tool_calls_concurrent(agent, assistant_message, messages: list, effe
     batch = _ConcurrentBatch(agent, messages, effective_task_id, parsed_calls, timeout_s)
     agent._current_tool = tool_names_str
     agent._touch_activity(f"executing {num_tools} tools concurrently: {tool_names_str}")
+
 
     spinner = _start_quiet_tool_spinner(agent, "", {}, label=f"⚡ running {num_tools} tools concurrently")
     try:
--- a/agent/tool_executor.py
+++ b/agent/tool_executor.py
@@ -53,6 +53,7 @@ from tools.tool_result_storage import (
     extract_persisted_path,
 )
 from tools.budget_config import BudgetConfig, DEFAULT_BUDGET, budget_for_context_window
+from agent.tool_result_validator import validate_tool_result, get_result_preview
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


Enough1122:

Better than a pure-additive change: this one is wired in — `_execute` in `agent/tool_executor.py` calls `validate_tool_result` after each tool execution, logs on failure, and records to `middleware_trace`. The lenient stance (unknown tools always pass) is the right default, and the tests are decent for the pure function. A few concerns:

1. **`middleware_trace` isn't defined anywhere in the diff — and the broad `except Exception` would hide it if it's missing.** The added block references `middleware_trace.append(...)` but the diff shows no definition; if it isn't a pre-existing local in `_execute`'s scope, that append raises `NameError`, which the surrounding `except Exception as _val_err` swallows and logs only at debug. That would silently kill the entire feature (and mask the bug). Please confirm `middleware_trace` is a defined local there — and more importantly, consider narrowing the catch so a genuine bug in the validator/record step is visible rather than swallowed as "best-effort".

2. **The `except Exception` + `logger.debug` pairing defeats the feature's own purpose.** The whole point is observability ("catches malformed results early"), yet if the validator itself raises, the incident is invisible at normal log levels. Any regression in the validator (import error, signature change, a new result type it doesn't handle) degrades silently. At minimum log at WARNING when the validator itself fails, not debug.

3. **Module docstring overclaims: this is observability, not prevention.** "Prevents downstream hallucinations" / "catches malformed results early" implies the bad result is stopped; in practice the original result always passes through (which the `_execute` comment correctly states). If the intent is prevention, wire the invalid path to actually do something (strip/flag the result before it reaches the model, or surface it to the UI). If it's telemetry only, the docstring and the `middleware_trace` name should say "tracks", not "prevents" — otherwise the stated contract and the shipped behavior drift apart.

4. **False-positive risk: any dict containing an `"error"` key is invalid for `web_search`/`web_extract`.** Legitimate structured results (MCP tools, providers that return an error-shaped body the model should actually see and reason about) would be flagged and warned on every call. Combined with `_validate_api_tool_result`'s dict branch being effectively a no-op ("has results/data/items → valid", "other keys → probably ok"), the api validation catches almost nothing real while still being able to misfire on an `"error"` key. If error-keyed results are data (the terminal validator already says "error output is data, not a failure" for exactly this reason), the web validator should be consistent and not flag them.

5. **No test exercises the wiring in `_execute`.** The suite only calls `validate_tool_result` directly. The integration behavior — the warning is emitted and `middleware_trace` gets an entry when a tool returns a bad shape — is untested, and a regression in the record step (like #1) wouldn't be caught. A test that runs a real tool through `_execute` with a malformed result and asserts the trace entry / log line would close that.

Minor: `_validate_file_tool_result` and `_validate_api_tool_result` use f-string `logger.warning(...)` instead of lazy `%s` formatting — the file's own code uses both styles; pick one (the module's other calls use `%s`).

The heuristic surface is small and honest; the main asks are: prove `middleware_trace` exists in scope, make validator-internal failures visible, and align the docstring with the actual (non-blocking) behavior.

thewulf7:
@Enough1122 review again please

Enough1122:

Re-reviewed at head `0976fead3fa60f43f7651585a2873ecb26be8680` (2 commits). Thanks for the request — here is where it stands against the points raised earlier.

**Resolved since the previous review**

- `middleware_trace` is a genuine local in the enclosing scope: it is bound as the loop variable at `agent/tool_executor.py:1224` (`for i, (tc, name, args, middleware_trace, block_result, _scope_block) in enumerate(parsed_calls)`) and reassigned at `:1403` (`middleware_trace = managed.middleware_trace`). The `NameError` risk I flagged is not real — verified against the base revision.
- The validator's own exception is now logged at WARNING rather than DEBUG, so a regression inside the validator no longer disappears at default log levels.

The wiring itself is also correct on one subtle point worth confirming: the `middleware_trace.append(...)` at the new block runs *before* `_emit_terminal_post_tool_call(..., middleware_trace=list(middleware_trace))` (base `:1448-1456`), so the validation entry really does reach the terminal `post_tool_call` payload rather than being appended to a list that was already copied.

**Remaining findings**

1. **Dead code: `ToolResultValidationError` and `get_result_preview` are never used.** In `agent/tool_result_validator.py`, the exception class at `:64` is defined and never raised or caught anywhere, and `get_result_preview` at `:165` has no caller — I confirmed both identifiers occur exactly once each in the file at head. Meanwhile the executor wiring builds its own preview with `repr(result)[:200]` instead of calling the helper the module ships. Either delete both, or have the wiring call `get_result_preview(result)` so truncation/sanitisation lives in one place. As written the module advertises two APIs nobody consumes. *(Non-blocking)*

2. **`TestWiringContract` tests a mirror of the wiring, not the wiring.** `tests/agent/test_tool_result_validator.py:364-376` re-implements the executor block as `_simulate_executor_wiring`, with its own local `trace: list = []`, then calls `validate_tool_result` directly. If the real block in `tool_executor.py` is deleted or edited, these tests still pass. They verify the validator's return contract, which is already covered above them — they do not verify integration. A test that runs a real tool call through `_execute` and asserts on the resulting `middleware_trace` would be the actual proof. *(Non-blocking, but the "Wiring contract" name currently overstates what is tested.)*

3. **Validation runs on the synthetic error string.** The new block sits after the `except Exception as tool_error:` handler (base `:1443-1445`), which sets `result = f"Error executing tool '{function_name}': {tool_error}"`. That string is then fed to the validator as though it were a tool result. It happens to pass (it is a `str`), so nothing breaks today, but validating a string the executor itself fabricated is not what "catch malformed results" means. Consider skipping validation when the tool raised. *(Non-blocking)*

4. **`blocked` / `dispatched` are not excluded.** `blocked` and `dispatched` are set at base `:1404-1405`, and the new block runs unconditionally for both. When a middleware dispatches the call elsewhere, `result` is the dispatcher's return value rather than a tool result, so checking it against the `read_file` / `terminal` / `search_files` rules can emit spurious warnings for shapes those rules were never written for. The very next statement already uses exactly this guard — `if not blocked and not dispatched:` at base `:1447`. Reusing it here keeps the semantics consistent. *(Non-blocking)*

5. **Per-call import inside the hot path.** `from agent.tool_result_validator import validate_tool_result` is executed inside `_execute`, i.e. on every tool call. Module caching makes this cheap, but the file imports everything else at module scope (`agent/tool_executor.py:13-54`), so a module-level import would match local convention. *(Non-blocking)*

**Verdict: LGTM** — the two functional concerns from the earlier review are resolved and the remaining items are hygiene. Items 1 and 2 are the ones I would actually act on before merge: each ships an API or a test whose stated purpose is not realised by the code.

thewulf7:
@Enough1122 re-review please

thewulf7:
@teknium1 hey could you review this one please?

PR #97224, opened 2026-08-28
feat(agent): checkpoint strategy and streaming heuristics for tool results

thewulf7:
## What does this PR do?

Adds two independent modules that improve agent robustness and responsiveness. This update addresses all points raised in the review.

---

### 1. CheckpointStrategy — *when* to persist agent state

A declarative decision layer that tells the agent loop *when* to checkpoint. Returns a `bool` — it does not call the checkpointer itself, so different loop implementations can plug in their own persistence.

Four strategies:

| Strategy | Behaviour |
|---|---|
| `NEVER` | No checkpoints (lightweight sessions) |
| `ALL` | After every tool call (maximum durability) |
| `RISKY` | After destructive tools only (write, patch, terminal, …) |
| `SMART` | Destructive tools + any error-result tool; skips successful reads |

**How it's enabled**: the strategy hooks into the *existing* `agent._checkpoint_mgr.enabled` gate in `tool_executor.py`. If you have `checkpoints: enabled: true` in `config.yaml`, `CheckpointStrategy.SMART` activates automatically — no new config key, zero overhead for users who don't use checkpoints.

**Integration**: `should_checkpoint()` is now called in `tool_executor.py` immediately after tool execution. It delegates to the existing `tools.checkpoint_manager.CheckpointManager.ensure_checkpoint()` — this PR extends that infrastructure rather than duplicating it.

---

### 2. Streaming heuristics — *whether* to stream a result progressively

`should_stream_tool(tool_name, result)` answers: is this result large enough that streaming is meaningfully better than one batch?

Rules (size/type-based, no hardcoded tool name list):
- Any tool: string > 1 KB → stream
- Any tool: list with ≥ 2 items → stream
- `None` → never stream

`StreamedToolResult` chunks text into configurable pieces. `stream_text_async()` now delegates to `stream_text()` so chunking logic lives in exactly one place.

---

## Review fixes (v2)

| Issue | Fix |
|---|---|
| Neither module wired into any code path | `should_checkpoint()` wired into `tool_executor.py` post-execution path, gated by existing `_checkpoint_mgr.enabled` |
| Duplicates existing checkpoint infra | Now extends `tools.checkpoint_manager.CheckpointManager` rather than introducing a parallel bookkeeping layer |
| `mcp_*` prefix match broken | Replaced literal set entry with `_is_mcp_tool()` prefix check; `get_checkpoint_label()` produces correct labels for `mcp_<server>_<tool>` names |
| SMART docstring contradicts impl | Rewritten: clearly states destructive ops + error results → checkpoint; successful API calls → no checkpoint |
| Sync/async streaming duplicate logic | `stream_text_async()` now delegates to `stream_text()` |
| `known_tools` hardcoded list anti-pattern | `should_stream_tool()` now gates on result size/type only — works for any current or future tool |

## Type of Change

- [x] 🐛 Bug fix (mcp_* prefix, async duplication)

## Changes Made

- `agent/checkpoint_strategy.py` — fixed MCP prefix matching, corrected SMART docstring, `%s`-style logging throughout
- `agent/streaming_results.py` — `stream_text_async` delegates to `stream_text`; `should_stream_tool` size/type-based only
- `agent/tool_executor.py` — wired `should_checkpoint()` into post-execution path behind existing `_checkpoint_mgr.enabled` guard
- `tests/agent/test_checkpoint_streaming.py` — 34 tests (added: async parity, MCP label, size-based streaming for unknown tools)

## How to Test

```bash
uv run --extra dev python3 -m pytest tests/agent/test_checkpoint_streaming.py -v
# → 34 passed
```

To test the wired path, enable checkpoints in `config.yaml`:

```yaml
checkpoints:
  enabled: true
```

Then run an agent session that calls `write_file` or `patch` — a checkpoint will be taken after each mutation.

### Code

- [x] I searched for existing PRs — no existing PR covers checkpoint strategy or streaming heuristics
- [x] My PR addresses all review comments from v1
- [x] I've run `pytest tests/agent/test_checkpoint_streaming.py -v` — 34 passed
- [x] Tested on: macOS 15 (Apple Silicon)

- [x] Module docstrings explain design decisions
- [x] No new config keys — activates via existing `checkpoints.enabled` setting
- [x] AGENTS.md unchanged — no architecture changes

### DIFF
--- a/agent/agent_init.py
+++ b/agent/agent_init.py
@@ -334,6 +334,7 @@ def _refuse_checkpoint_required_on_codex_app_server(
         )
 
 
+
 def _parse_config_int(raw: Any, default: int) -> int:
     """Strict int coercion: rejects bool (YAML ``true`` → 1) and fractional floats."""
     if isinstance(raw, bool):
@@ -1142,13 +1143,36 @@ def _init_session_state(agent, session_id, session_db, parent_session_id, reason
 
     # Filesystem checkpoint manager (transparent — not a tool)
     from tools.checkpoint_manager import CheckpointManager
+    from agent.checkpoint_strategy import CheckpointStrategy
     agent._checkpoint_mgr = CheckpointManager(
         enabled=checkpoints_enabled, max_snapshots=checkpoint_max_snapshots,
         max_total_size_mb=checkpoint_max_total_size_mb,
         max_file_size_mb=checkpoint_max_file_size_mb,
     )
 
-    agent._session_db = session_db  # optional SQLite store (CLI/gateway-provided)
+    # Strategy used by the post-execution checkpoint hook in tool_executor.
+    # Defaults to SMART (checkpoint after mutations or error results).
+    # Can be overridden via the checkpoint_post_strategy init param.
+    try:
+        agent._checkpoint_post_strategy = CheckpointStrategy(checkpoint_post_strategy)
+    except ValueError:
+        logger.warning(
+            "Unknown checkpoint_post_strategy %r; falling back to SMART",
+            checkpoint_post_strategy,
+        )
+        agent._checkpoint_post_strategy = CheckpointStrategy.SMART
+    
+    # SQLite session store (optional -- provided by CLI or gateway)
+    agent._session_db = session_db
+    # Whether close() must also close that handle. Default False: a
+    # caller-supplied session_db is almost always the SHARED launch handle,
+    # which outlives every agent and must never be closed here. Callers that
+    # hand over a DEDICATED handle (the gateway's per-profile state.db opens)
+    # set this True at the point ownership transfers, so teardown releases the
+    # sqlite fds and the token-writer thread instead of leaking them for the
+    # life of the process. Also set True on the lazy self-open in
+    # _get_session_db_for_recall, where nothing else holds a reference.
+    agent._owns_session_db = False
     agent._parent_session_id = parent_session_id
     agent._session_init_model_config = {
         "max_iterations": agent.max_iterations,
--- a/agent/checkpoint_strategy.py
+++ b/agent/checkpoint_strategy.py
@@ -1,19 +1,31 @@
-"""Checkpoint strategy middleware.
+"""Checkpoint strategy: decide *when* to trigger agent state persistence.
 
-Decides when to save agent state for replay/recovery.
+This is a pure decision layer.  It returns ``True``/``False`` and never
+calls the checkpointer directly, so different loop implementations can
+plug in their own persistence backend (e.g. the existing
+``tools.checkpoint_manager.CheckpointManager``).
 
-Strategies:
-- checkpoint_never: no checkpoints
-- checkpoint_all: after every tool call
-- checkpoint_risky: after destructive operations (write, delete, API calls)
-- checkpoint_smart: after uncertain operations (API calls, LLM reasoning changes)
+Four strategies:
+
+| Strategy | Triggers checkpoint                                 |
+|----------|-----------------------------------------------------|
+| NEVER    | Never                                               |
+| ALL      | After every tool call                               |
+| RISKY    | After destructive/mutating operations only          |
+| SMART    | After destructive ops OR error results (see below)  |
+
+SMART rationale: checkpoint after any state-changing op (write, patch,
+terminal, move/delete) and after any tool that returns an error dict —
+the agent may be in an inconsistent state and worth snapshotting before
+a retry loop starts.  Successful read-only API calls do *not* trigger a
+checkpoint; the result is ephemeral data and nothing has changed.
 """
 
 from __future__ import annotations
 
 import logging
 from enum import Enum
-from typing import Any, Dict, Optional, List
+from typing import Any, Dict, List, Optional
 
 logger = logging.getLogger(__name__)
 
@@ -21,27 +33,27 @@ logger = logging.getLogger(__name__)
 class CheckpointStrategy(Enum):
     """When to checkpoint agent state."""
 
-    NEVER = "never"  # No checkpoints
-    ALL = "all"  # After every tool call
-    RISKY = "risky"  # Only after destructive operations
-    SMART = "smart"  # After API calls, file mutations, uncertain outcomes
+    NEVER = "never"   # No checkpoints
+    ALL = "all"       # After every tool call
+    RISKY = "risky"   # Only after destructive/mutating operations
+    SMART = "smart"   # After destructive ops or error results
 
 
-# Tool categories for checkpoint decisions
+# ---------------------------------------------------------------------------
+# Tool category sets
+# ---------------------------------------------------------------------------
+
 _DESTRUCTIVE_TOOLS = {
     "write_file",
     "patch",
-    "terminal",  # Can modify files
+    "terminal",   # can modify files
     "delete_file",
     "move_file",
 }
 
-_API_TOOLS = {
-    "web_search",
-    "web_extract",
-    "mail",
-    "mcp_*",  # MCP tools
-}
+# MCP tools use the naming convention  mcp_<server_id>_<tool_name>
+# Match by prefix rather than a literal "mcp_*" entry.
+_MCP_PREFIX = "mcp_"
 
 _UNCERTAIN_TOOLS = {
     "execute_code",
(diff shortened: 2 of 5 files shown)


Enough1122:

The biggest issue with this PR is structural: **neither module is wired into any existing code path.** The diff adds `agent/checkpoint_strategy.py`, `agent/streaming_results.py`, and unit tests only — nothing in `run_agent.py`, `model_tools.py`, the gateway, or the TUI calls `should_checkpoint`, `CheckpointManager`, `should_stream_tool`, or `stream_tool_result`. As shipped, this is speculative infrastructure with no consumer, which the repo's contribution rubric explicitly rejects ("adding a hook is easy; removing one after plugins depend on it is hard. A hook is NOT speculative if a contributor has a real, stated use case"). The tests pass because they only test the modules against themselves — there's no proof the feature does anything in a real agent loop. Either wire it into the actual tool-call loop (checkpoint after each tool result, stream large tool results through the gateway/TUI) with an E2E test, or hold off on shipping the surface.

A few concrete issues in the code itself:

1. **This duplicates the existing checkpoints system rather than extending it.** Hermes already has a `checkpoints` config section and checkpoint plumbing in the agent (checkpoints config is a first-class `AIAgent` parameter). This PR defines a parallel `CheckpointManager` + `CheckpointStrategy` enum with its own history/labels, unaware of that infrastructure. Per the repo's "Extend, don't duplicate" rule, this should integrate with the existing checkpoint config/session-store — e.g. a strategy that decides *when* the existing checkpoint mechanism fires — not introduce a second, disconnected bookkeeping layer that nothing reads.

2. **The `mcp_*` prefix matching can never match a real tool.** `_API_TOOLS = {"...", "mcp_*"}` and `get_checkpoint_label` checks `tool_name.startswith(f"{api}:")`, producing `"mcp_*:"` as the prefix — no real MCP tool is named `mcp_*:something`. So the `after_<name>_call` label branch for MCP tools is dead. If the intent is "any tool in an mcp toolset", match on the toolset/schema source, not a literal `mcp_*` string.

3. **The SMART strategy's docstring and implementation disagree.** Module/class docs say "checkpoint_smart: after uncertain operations (API calls, LLM reasoning changes)", but the implementation's step 3 explicitly returns False for successful API calls ("Successful API calls do NOT trigger a checkpoint"). One of the two is wrong — if the design intent is "uncertain outcomes", an error-dict check plus destructive ops is reasonable, but then the docs should say so; if "after API calls" is intended, step 3 contradicts it. Either way it reads as if the behavior drifted from the stated contract.

4. **Sync/async streaming duplicate logic.** `stream_text_async` re-implements `stream_text` rather than delegating (e.g. an async iterator that yields from the same chunking). The non-string JSON-serialization branch exists twice. If chunking or the serialization fallback evolves, the two drift. A single chunking helper (sync), with the async variant `async for`-yielding from it, would keep them in lockstep.

5. **`should_stream_tool`'s `known_tools` is a frozen list of tool names** — the exact shape the repo warns against as a change-detector (any toolset renames/additions silently change behavior with no test signal). If streaming is meant to apply broadly, gate on result size/type generically; if it's tool-specific, derive the set from the registry rather than hardcoding.

The unit tests are fine as far as they go, but they assert the heuristics in isolation — the missing piece is any test that exercises these against the real tool-result pipeline (a checkpoint actually landing in the session store, a large result actually streaming through the gateway). Nice idea; it needs a consumer and an integration test to be a feature rather than dead code.

thewulf7:
@Enough1122 review again please

Enough1122:

Re-reviewed at head `4efccf922e15d413746f503fd966a0f5de1a736b` (2 commits). Good progress on the structural point from the earlier review — the checkpoint half is now wired, and `should_stream_tool` no longer gates on a hard-coded tool-name list. What is left is mostly about the wiring landing correctly.

**Resolved since the previous review**

- No longer speculative on the checkpoint side: `agent/tool_executor.py` now calls `should_checkpoint(...)` and `agent._checkpoint_mgr.ensure_checkpoint(cwd, label)` from inside `_execute`, so the decision layer has a real consumer.
- `should_stream_tool` (`agent/streaming_results.py:254-275`) now decides on result size and type rather than a tool-name allowlist — the anti-pattern flagged earlier is gone.
- `stream_text_async` delegates to `stream_text` (`:233-240`), so chunking and serialisation exist in one place.

**Blocking: the new post-execution checkpoint path conflicts with the pre-execution one that already exists**

1. **It resolves the working directory differently, and the existing code explicitly warns against the way this does it.** There is already a pre-execution checkpoint at `agent/tool_executor.py:1073-1082` → `_ensure_file_checkpoint()` (`:85-101`), which resolves the path first: `_resolve_path_for_task(file_path, effective_task_id)` → `agent._checkpoint_mgr.get_working_dir_for_path(str(resolved_path))` → `ensure_checkpoint(work_dir, f"before {function_name}")`. The comment at `:93-96` states the reason outright: *"File tools resolve relative paths against the task's live/session cwd, which can differ from the Hermes process cwd (notably in Docker)."*

   The new block instead uses `cwd = function_args.get("workdir") or os.getenv("TERMINAL_CWD", os.getcwd())`. `"workdir"` is a terminal-tool argument; `write_file` and `patch` carry `path`, so for those two this expression always falls through to `os.getcwd()` — the process cwd the existing comment warns about. For a `write_file` in Docker, the pre-checkpoint therefore lands in the correct task-resolved project root and the new post-checkpoint lands somewhere else, which can register a second spurious project root. Please route through the same `_resolve_path_for_task` / `get_working_dir_for_path` pipeline as `_ensure_file_checkpoint`.

2. **It removes the destructive-command gate for `terminal`, so read-only commands now checkpoint.** The existing terminal rule at `:1084-1095` checkpoints only when `_is_destructive_command(command)` is true. The new strategy puts `"terminal"` in `_DESTRUCTIVE_TOOLS` (`agent/checkpoint_strategy.py:52-58`) unconditionally, so `ls`, `git status`, `cat`, `grep` and every other read-only command now trigger a filesystem snapshot after every call. The module docstring's own SMART rationale is "checkpoint after any state-changing op" — a bare `ls` is not one. Either consult the command before deciding, or drop `terminal` from `_DESTRUCTIVE_TOOLS` and let the call site pass the destructive-command result in.

**Remaining findings**

3. **`agent/streaming_results.py` is still entirely unwired.** `should_stream_tool`, `stream_tool_result` and `StreamedToolResult` have no caller anywhere in the diff outside `tests/agent/test_checkpoint_streaming.py`. This is the same "no consumer" concern as before, now applying to half the PR rather than all of it. `stream_multimodal` (`:242-251`) is additionally untested. If streaming is not part of this change, splitting it into a follow-up would leave the checkpoint half mergeable.

4. **`CheckpointStrategy.SMART` is hard-coded at the only call site**, so `NEVER`, `ALL` and `RISKY` are unreachable in production, and `CheckpointManager` (`agent/checkpoint_strategy.py:133-171`) is never instantiated outside tests — the wiring imports only `should_checkpoint`, `get_checkpoint_label` and `CheckpointStrategy`. The `checkpoints_taken` / `checkpoint_history` bookkeeping therefore records nothing in a real run. A four-way enum with no way to select among the values is configuration surface without a knob behind it; add the config option or trim the enum to what is shipped.

5. **Failures are swallowed at DEBUG.** The block ends with `except Exception as _cp_err: logger.debug(...)`. Combined with items 1 and 2, a wrong-directory or failing `ensure_checkpoint` is invisible at normal log levels — the same concern raised on #97222, where it was addressed by moving to WARNING. Given this hook can fire on every tool call, a persistent failure here would be silent. At minimum log at WARNING.

6. **This introduces concurrent `ensure_checkpoint` calls.** The existing checkpoints all happen on the main thread before dispatch (`:1073-1095`); the new block runs inside `_execute`'s worker, so a parallel batch of N destructive tools now issues N concurrent snapshots. If `ensure_checkpoint` or its underlying snapshot is not internally synchronised, this is a new race. Worth confirming explicitly.

7. **Minor:** `_UNCERTAIN_TOOLS` (`agent/checkpoint_strategy.py:64-67`) is defined and never read. `_STREAM_TEXT_MIN_BYTES` is documented as bytes (`:199-200`) but compared against `len(result)`, which is characters — for non-ASCII output 1024 characters is up to ~3 KB. The tests use `asyncio.get_event_loop().run_until_complete` (`tests/agent/test_checkpoint_streaming.py:490`), deprecated since 3.12 and an error with no running loop; `asyncio.run` is the current form. *(All non-blocking)*

**Verdict: Request changes** — items 1 and 2 are behavioural regressions against checkpoint logic that already exists in the file, not stylistic preferences. Item 3 is the one to decide on scope: either wire streaming or split it out.

thewulf7:
@Enough1122 please review again

Enough1122:

Re-reviewed incremental commit 9318d80 against the 09-01 pass: terminal correctly dropped from _DESTRUCTIVE_TOOLS with the command-level _is_destructive_command gate (read-only commands no longer snapshot) plus a pinning test; file-tool cwd now routes through _resolve_path_for_task (fixes Docker/task-isolated misplacement); configurable checkpoint_post_strategy with SMART fallback; hook failures promoted debug to warning. Non-blocking: audit that no shared-handle caller ever sets _owns_session_db=True, and confirm _is_destructive_command heuristics cover piped/compound destructive commands. Looks good.

thewulf7:
@teknium1 hey could you review this one please?

PR #97227, opened 2026-08-28
fix(delegation): resolve reliable worktree repo anchors

### LINKED ISSUE #97209: delegation.worktree_isolation silently no-ops when the parent session's recorded cwd is not a git repo (persistent/owner sessions)

fangliquanflq:
## Summary

- add `delegation.worktree_repo_root` as an explicit absolute repository anchor for persistent sessions
- resolve every available workspace candidate when the recorded session cwd is stale or outside a repository
- surface worktree isolation degradation in the child context, application logs, and the parent delegation result
- document anchor precedence and cover it with real git repositories plus delegation-result behavior tests

## Testing

- `scripts/run_tests.sh tests/tools/test_subagent_worktree.py -q`
- `scripts/run_tests.sh tests/tools/test_delegate.py -k worktree_degradation_is_visible_to_child_and_parent -q`

Closes #97209

### DIFF
--- a/hermes_cli/config_defaults.py
+++ b/hermes_cli/config_defaults.py
@@ -2062,6 +2062,9 @@ DEFAULT_CONFIG = {
                                        # delegation units. New async dispatches beyond the cap
                                        # fall back to synchronous execution. Floor of 1, no ceiling.
                                        # (Replaces the deprecated max_async_children.)
+        "worktree_isolation": False,  # opt-in local git worktree per child
+        "worktree_repo_root": "",  # optional absolute checkout anchor; useful for
+                                    # persistent sessions whose recorded cwd is stale
         # Orchestrator role controls (see tools/delegate_tool.py:_get_max_spawn_depth
         # and _get_orchestrator_enabled).  Floored at 1, no upper ceiling —
         # raise deliberately, each level multiplies API cost.
--- a/tests/tools/test_delegate.py
+++ b/tests/tools/test_delegate.py
@@ -517,6 +517,46 @@ class TestToolNamePreservation(unittest.TestCase):
 class TestDelegateObservability(unittest.TestCase):
     """Tests for enriched metadata returned by _run_single_child."""
 
+    def test_worktree_degradation_is_visible_to_child_and_parent(self):
+        parent = _make_mock_parent(depth=0)
+        captured = {}
+
+        with (
+            patch("run_agent.AIAgent") as MockAgent,
+            patch("tools.delegate_tool._get_worktree_isolation", return_value=True),
+            patch("tools.subagent_worktree.local_backend_active", return_value=True),
+            patch(
+                "tools.subagent_worktree.resolve_worktree_anchor",
+                return_value=(None, "no usable repository; using shared workspace"),
+            ),
+        ):
+            mock_child = MagicMock()
+            mock_child.model = "claude-sonnet-4-6"
+            mock_child.session_prompt_tokens = 0
+            mock_child.session_completion_tokens = 0
+
+            def run_child(user_message, **_kwargs):
+                captured["message"] = user_message
+                return {
+                    "final_response": "done",
+                    "completed": True,
+                    "interrupted": False,
+                    "api_calls": 1,
+                    "messages": [],
+                }
+
+            mock_child.run_conversation.side_effect = run_child
+            MockAgent.return_value = mock_child
+
+            result = json.loads(
+                delegate_task(goal="Test warning visibility", parent_agent=parent)
+            )
+
+        entry = result["results"][0]
+        self.assertIn("shared workspace", entry["worktree_warning"])
+        self.assertIn("[WORKTREE ISOLATION WARNING]", captured["message"])
+        self.assertIn(entry["worktree_warning"], captured["message"])
+
     def test_observability_fields_present(self):
         """Completed child should return tool_trace, tokens, model, exit_reason."""
         parent = _make_mock_parent(depth=0)
--- a/tests/tools/test_subagent_worktree.py
+++ b/tests/tools/test_subagent_worktree.py
@@ -11,6 +11,7 @@ import tempfile
 import shutil
 import unittest
 from pathlib import Path
+from types import SimpleNamespace
 from unittest import mock
 
 sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..")))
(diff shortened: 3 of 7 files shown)


Enough1122:

**Overall:** Reliable worktree anchor resolution with stale-cwd and corrupted-index handling.

**What it does**
- `tools/subagent_worktree.py` resolves `worktree_repo_root` from config (`delegation.worktree_repo_root`) as explicit anchor when parent `cwd` is non-git (persistent sessions with stale `recorded cwd`). Handles linked worktrees where `.git` is a file (`gitdir: …`) via `rev-parse --git-path index` fallback and correctly writes corrupted index to resolve `index` file.
- `tools/delegate_tool.py` threads `worktree_repo_root` through delegation; `hermes_cli/config_defaults.py` defaults `worktree_isolation:false`, `worktree_repo_root:""`.
- Tests cover `_make_repo` parents creation, `_break_git_index` for file vs directory git, `DelegationConfigGateTests` for isolation default/off.

**Non-blocking notes**
- `_break_git_index` now accepts `Path|str` — good, but caller in `test_delegate_tool_fallback_uses_the_shared_factory` passes `wt` as path; ensure `Path(wt).resolve()` handles worktree with external gitdir correctly.
- `worktree_repo_root` anchor overlaps #97226/#97228 — three PRs implementing same feature; unify to one to avoid drift. This PR's `resolve_worktree_anchor` logic looks most complete.

Solid worktree reliability fix; needs coordination with sibling PRs.

fangliquanflq:
Thanks for the review. I checked both notes against the current head:

- No code change is needed for the linked-worktree index path. `_break_git_index` is typed as `Path`, and its only callers pass `Path` objects. It obtains the actual worktree git directory with `git rev-parse --git-dir`, using an absolute external gitdir directly and resolving only relative results against the worktree. `test_delegate_tool_fallback_uses_the_shared_factory` does not call `_break_git_index`; it exercises `unproven_worktree_payload` instead. The real linked-worktree corruption tests cover the external-gitdir path.
- The overlap with #97226 and #97228 is real coordination information, but it does not identify a defect in this branch. All three target #97209; this PR keeps the more complete candidate resolution and exposes degradation to the child context, logs, and parent result. The duplicate PRs should be coordinated rather than combining their implementations here.

Verified with:
- `scripts/run_tests.sh tests/tools/test_subagent_worktree.py -q` — 31 passed
- `scripts/run_tests.sh tests/tools/test_delegate.py -k worktree_degradation_is_visible_to_child_and_parent -q` — 1 passed

PR #97061, opened 2026-08-28
fix(agent): quarantine same-workspace delegated mutations until stale child exits

astraltrekkin:
## What does this PR do?

Adds an owner-aware workspace mutation admission fence for delegated children. When a delegated child times out or is cancelled but stays live after the existing tool-interrupt grace window, later known built-in mutations in the same trusted local workspace are now denied until every stale owner actually completes. Reads, unknown plugin/MCP tools, mutations in other workspace domains, and non-delegated callers remain unaffected. The quarantine is released only by real completion (the child's future finishing), not by elapsed time alone.

This is the small, explicit lifecycle contract the issue asks for, rather than an unconditional global lock: each delegated child is bound to its workspace domain at dispatch; on genuine timeout/cancel its mutation gate closes and interruption is requested; after the existing bounded grace the domain is quarantined only if the child is still live; at the central execution boundary only known built-in mutating tools for later delegated work in that quarantined domain are denied.

## Related Issue

Fixes #95874

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `agent/workspace_mutation_fence.py` (new): owner registry, mutation gate, grace-then-quarantine-if-still-live, `deny_delegated_mutation` admission decision, and `release_owner` on real completion. Fencing is scoped to `FILE_MUTATING_TOOL_NAMES` (`write_file`, `patch`) and delegated-child contexts only.
- `tools/delegate_tool.py`: in `_run_single_child`, bind the child as owner of its resolved workspace domain at dispatch, set liveness to `not future.done()`, register a done-callback that releases the owner on real completion, and call `mark_timeout_or_cancel` on the timeout branch. The batch abandon path also calls `mark_timeout_or_cancel` for abandoned children. The existing timeout/cancellation result returned to the caller is unchanged.
- `model_tools.py`: in `handle_function_call`, add a pre-dispatch admission check (`deny_delegated_mutation`) between the plugin-block check and the ACP edit-approval check; on deny it returns a `tool_error` with `error_type="workspace_quarantine"` and emits the post-tool hook with `status="blocked"`.
[...]

PR #96255, opened 2026-08-27
Make Obsidian the authoritative persistent memory source

omxhub:
## Root cause

`memory.authoritative_source: obsidian` was **inert** — no runtime code consumed it. The built-in `~/.hermes/memories` store remained the system-prompt source of truth, and the Obsidian plugin only *mirrored* writes. Empirical tests confirmed: a vault-only memory was only recalled via the plugin's additive block, and in a conflict (built-in = A, vault = B) the built-in value won.

## Fix

`get_memory_dir()` now resolves to `<vault>/Memory` when `memory.authoritative_source: obsidian` is set and the vault is reachable. Since every reader (built-in `MemoryStore`, delegates, context-compression, learning graph, `hermes backup`) goes through `get_memory_dir()`, Obsidian becomes the single canonical persistent-memory location — no fragmenting of state across two stores.

## Safe migration & fallback

- **One-time merge:** on the first flip, entries already in the classic `~/.hermes/memories` store are merged (union + deduplication, near-duplicate collapse keeping the classic wording) into the vault so no memory is dropped. Guarded by a `.migrated` sentinel and re-runnable-safe.
- **Safe fallback:** if the vault is unconfigured or unreachable, `get_memory_dir()` falls back to the classic store — the agent never breaks.
- Atomic writes preserved (`os.replace`).

## Compatibility

HADES, Mission Control, Hermes Bridge, the memory tools, and transient conversation context are all unaffected — none of them couple to the old hardcoded `memories/` path; they read through `get_memory_dir()`.

## Verification

- Memory tests: **100 passed**
- Obsidian lifecycle: **14/14 passed** (real vault, post-restart)
- Real Hermes restart verified (injected memory now sourced from vault)
- Direct Obsidian mutation verified (create / recall / edit / recall-updated)
- Conflict / source-of-truth behavior verified
- Unrelated Vault files verified untouched (`Welcome.md` mtime unchanged)

## Files changed

- `tools/memory_tool.py` (+137 / −2): `get_memory_dir()` honors `authoritative_source: obsidian`; one-time fail-safe merge; safe fallback.

(Note: `plugins/obsidian/__init__.py` docstring was corrected for accuracy but is outside this repo and not part of the diff.)

PR #97231, opened 2026-08-28
feat: load one global policy across profiles

kbj2043:
## Summary
- load a single root-level `memories/GLOBAL.md` for every Hermes profile
- inject the policy as a separate frozen, sanitized system-prompt block
- keep profile `MEMORY.md` and `USER.md` stores independent and writable only through existing targets
- add regression coverage for root/profile resolution, snapshot freezing, and prompt injection

## Design
`get_default_hermes_root()` normalizes the default and named-profile layouts, so named profiles resolve the same canonical policy path without copying policy text into each profile.

A missing `GLOBAL.md` is a no-op. Existing memory threat scanning applies before system-prompt injection, and the snapshot remains frozen for the session like the existing memory blocks.

## Tests
- `venv/bin/python -m pytest tests/tools/test_memory_tool.py tests/agent/test_system_prompt.py -q` — 80 passed
- `venv/bin/ruff check tools/memory_tool.py agent/system_prompt.py tests/tools/test_memory_tool.py tests/agent/test_system_prompt.py` — passed
- `git diff --check` — passed

The full `tests/agent` suite was attempted but exceeded the local execution timeout; the targeted suite passed.


andrexibiza (commented):
Reviewed exact head `5bb2955213e579d5ef50ed6c54c89becff658ee1` against live `main@306db2776c6b6f1acc85c31c4dabba3263f0e9fd`. Exact merge base is `c30ac90a92097058ddd6f9db3fa2e3182a7bfdcc`; this head is 2 commits ahead / 67 behind current main.

There is a lot to like here. `get_default_hermes_root()` is the right primitive for one root-level source shared by named/default profiles, the policy bytes go through the existing memory sanitization path, and the second commit correctly made the GLOBAL renderer a distinct policy block instead of pretending it is personal MEMORY. Extending the policy to focused children is also the right architectural question to ask. I do see two correctness blockers in the exact current shape, plus landing/ownership gates.

### P1 — main-agent GLOBAL policy still disappears when both personal built-in stores are disabled

The new `load_global_policy_block()` is independent, but `agent/system_prompt.py::build_system_prompt_parts()` does not call it. Main-agent injection is still nested under:

```py
if agent._memory_store:
    global_policy_block = agent._memory_store.format_for_system_prompt("global_policy")
```

On this exact head, `agent/agent_init.py` initializes `_memory_store = None` and only constructs a `MemoryStore` when at least one of `_memory_enabled` / `_user_profile_enabled` is true. The outer initialization path is also bypassed by `skip_memory=True` unless the memory toolset was explicitly requested.

So a profile can have a perfectly valid machine-wide `memories/GLOBAL.md` and receive **no global policy at all** simply because it disabled both writable personal stores. Focused delegated children now do receive the policy because `_build_child_system_prompt()` calls `load_global_policy_block()` directly, which makes the asymmetry even sharper.

That contradicts the PR's stated ownership model: GLOBAL is described as one machine-wide policy source while profile `MEMORY.md` / `USER.md` remain independent. A machine policy should not depend on whether either personal-memory feature happens to instantiate a store.

**Required fix:** give the GLOBAL snapshot its own initialization/assembly path (or otherwise load it independently of the personal-memory flags), and add a production-construction regression with both built-in stores disabled. Please pin `skip_memory=True` explicitly too: either GLOBAL survives those execution modes because it is policy rather than memory, or the public contract must say that it is intentionally suppressed there. The current accidental coupling should not decide that policy.

### P1 — parent and delegated child can execute under different GLOBAL policy versions in one session

The PR promises a frozen snapshot: mid-session file writes should not change the system prompt until the next session. Main agents follow that contract because `MemoryStore.load_from_disk()` snapshots `global_policy` into `_system_prompt_snapshot`.

The new delegation path does not inherit that frozen snapshot. Every call to `_build_child_system_prompt()` performs a fresh:

```py
from tools.memory_tool import load_global_policy_block
global_policy_block = load_global_policy_block()
```

That creates a concrete split-brain sequence:

1. Parent session starts with GLOBAL policy A and freezes A.
2. `GLOBAL.md` changes on disk to B while the parent session remains alive.
3. The parent still executes under A, by design.
4. A later `delegate_task` child fresh-reads B and executes under B.

The same authority chain now has two different machine policies, even though the feature's cache/freeze contract says policy changes take effect at the next session boundary. The new delegate regression only proves that a child sees the current file; it does not prove parent→child version coherence.

**Required fix:** bind the exact frozen policy snapshot/version to the parent session and pass that snapshot into child prompt construction (or establish a different explicit generation boundary that updates parent and descendants atomically). Add an adversarial regression that mutates `GLOBAL.md` after parent prompt construction and proves a child spawned from that parent receives the same frozen bytes as the parent. A fresh parent session should then pick up B.

### Ownership / merge-order graph

This branch is now materially behind current main and touches two heavily contested seams, so the landing order matters:

- Teknium's merged #97165 is upstream of this review and records the maintainer decision that profiles are not a hard isolation boundary; it also changed `agent/system_prompt.py` after this branch's merge base. A shared read-only machine policy is compatible with that decision, but this PR must restack on current main without resurrecting the retired cross-profile guard semantics.
- DavidMetcalfe's open #96709 is complementary, not duplicate work: it changes the same `MemoryStore` frozen-snapshot / `format_for_system_prompt()` seam for project-scoped **personal** MEMORY. GLOBAL should remain machine-scoped and unfiltered. Whichever lands second needs an explicit composition test rather than two independent render authorities.
- omxhub's open #96255 is also complementary: it changes `get_memory_dir()` so personal MEMORY/USER can be sourced from an authoritative Obsidian vault. This PR deliberately roots GLOBAL at `get_default_hermes_root()/memories/GLOBAL.md`; preserve that distinction in a combined regression so an alternate personal-memory backend cannot relocate the machine policy.
- The second commit also grows `tools/delegate_tool.py`, whose live fracture owner is #79928. That same godfile is concurrently modified by open #97061 (rainbowgore, stale-child mutation ownership) and #96500 (686f6c61, delegated-child ContextVar restoration). Those are adjacent/complementary rather than duplicates, but this is exactly the kind of same-file lifecycle surface that needs one restack/merge order. The GLOBAL policy wiring is small enough to live behind a bounded prompt/policy owner instead of adding another responsibility directly to the fracture target.

### Attribution and exact-object acceptance

Both commits are authored as `Lukas <kbj2043@gmail.com>`, and live `main@306db277...` has no `contributors/emails/kbj2043@gmail.com` mapping. Please add the canonical mapping to `kbj2043`; preserve the existing commit authorship rather than rewriting it.

The exact current head also has no hosted-green acceptance receipt. CI `33182755591`, Docker `33182754951`, and Nix `33182755034` all completed as `action_required`; the CI run contains zero jobs. The previous commit `b32fdaab...` likewise had only `action_required` workflow runs. The reported local tests are useful development evidence, but neither commit is currently release-green.

The root-normalization, dedicated GLOBAL header, sanitization, and explicit child-policy coverage are good progress. The remaining work is about making the word “global” mean one actual authority snapshot across **all** execution paths, not merely one pathname. Close those two generation/initialization holes, restack the shared seams, and this becomes a much stronger primitive. 🚀

kbj2043:
## Review feedback addressed

Implemented on current upstream `main` (base `a9e72f1b5`) through head `c69ecaa31`.

### P1: GLOBAL policy depended on personal memory

Fixed by snapshotting the machine-wide policy independently during `AIAgent` construction. `GLOBAL.md` is now injected even when both personal stores are disabled and when `skip_memory=True`; personal `MEMORY.md`/`USER.md` remain independently gated and writable only through the existing memory tool.

Added a real `AIAgent` regression covering disabled personal memory with a temporary `HERMES_HOME`.

### P1: parent/child policy split-brain

Fixed by accepting an optional frozen `global_policy_snapshot` on `AIAgent` and passing the parent snapshot to delegated children. The child prompt no longer adds a duplicate global block; the child’s normal system prompt owns the single injected copy. A child without a parent snapshot still reads a fresh snapshot at its own session boundary.

Added coverage for the constructor handoff and child prompt behavior. A temporary named-profile E2E also verified:

- parent and child retain policy A after `GLOBAL.md` changes to B;
- a fresh session reads B;
- parent and child each contain exactly one policy block;
- named profiles resolve the root-level canonical policy path.

### Validation

- `pytest tests/agent/test_system_prompt.py tests/agent/test_skip_memory_store_65429.py tests/tools/test_delegate.py tests/tools/test_memory_tool.py -q` — **159 passed**
- `ruff check ...` — **passed**
- `git diff --check` — **passed**
- temporary named-profile parent/child E2E — **passed**

The existing commit authorship was preserved. Hosted GitHub checks are still pending/not reported for this fork branch; local and E2E evidence above is not being presented as a hosted-green receipt.

andrexibiza (commented):
Re-reviewed Lukas / kbj2043's exact head `c69ecaa314e14282529a1df33c9b4bf3b5837d08`, including the nine-file diff and the constructor → prompt restore → delegation paths, against inspected `main@2237be355906fbe6065ce1815711eee52b2d646e`. Merge base: `a9e72f1b5843467b7cabd4d361b8088832912afe`.

### Previous findings: one corrected, one only partially closed

The independent initialization of `_global_policy_snapshot` and its injection outside `_memory_store` correct the disabled-personal-memory coupling from [my earlier review](https://github.com/NousResearch/hermes-agent/pull/97231#pullrequestreview-5052326330). Passing the explicit snapshot to the child constructor also corrects the **same live parent instance** case; preserving `""` rather than treating it as a fresh-read request is important here. The focused child prompt no longer duplicates the normal prompt's GLOBAL block.

**One P1 remains: the snapshot is bound to an agent instance, not the restored logical session.** The inline comment traces the ordinary resumed-session path: the new instance reads policy B, the persisted parent prompt restores policy A, and delegation forwards B. Deleting GLOBAL.md between turns produces the stronger variant: the parent retains A while its child receives no GLOBAL policy. This is a remaining case of the original parent/child coherence finding, not a request to undo the corrected handoff.

The closure condition is one session-owned snapshot, including explicit absence, restored together with the persisted prompt before descendants or prompt rebuilds can consume it. Add a durable regression using a real SessionDB, a reconstructed AIAgent for the **same session ID**, and the production child builder. Cover A→B, A→missing, and absent→B, alongside unchanged-policy, same-live-parent, and genuinely new-session controls. Check the actual parent/child prompt bytes and single-copy rendering, not just the constructor argument.

### Verification boundary

I traced the exact-head source and ran an isolated mechanism harness using transcribed snapshot/restore/handoff statements, harmless prompt/loader fixtures, and temporary filesystem/SQLite state. It exercised seven scenarios: four coherent controls and the three resumed-session counterexamples above. **That is not a full AIAgent integration run or a rerun of the repository suite.** The [159-test result and named-profile E2E in the author follow-up](https://github.com/NousResearch/hermes-agent/pull/97231#issuecomment-5454730711) remain author-reported evidence; they are not represented here as independently rerun or as hosted-green acceptance.

### Landing and exact-commit checks

The [comparison with inspected main](https://github.com/NousResearch/hermes-agent/compare/2237be355906fbe6065ce1815711eee52b2d646e...c69ecaa314e14282529a1df33c9b4bf3b5837d08) is **4 commits ahead / 6,655 behind**. GitHub currently reports `mergeable=false`; the August 28 restack is not a current-main landing receipt.

| Current PR commit | Retrieved PR-triggered workflow evidence |
| --- | --- |
| `15e69300537c0f50b40fe891635b8fb6e3969042` | No runs returned for this exact SHA. |
| `bd9ffb097a460a9eb439ed078aa037395277c70c` | No runs returned for this exact SHA. |
| `b32019b3c04b56b0f96f7ec577a52c0955aa496a` | [CI](https://github.com/NousResearch/hermes-agent/actions/runs/33187547839), [Docker](https://github.com/NousResearch/hermes-agent/actions/runs/33187547111), [Nix](https://github.com/NousResearch/hermes-agent/actions/runs/33187547124): all `action_required`. |
| `c69ecaa314e14282529a1df33c9b4bf3b5837d08` | [CI](https://github.com/NousResearch/hermes-agent/actions/runs/33187686315), [Docker](https://github.com/NousResearch/hermes-agent/actions/runs/33187685816), [Nix](https://github.com/NousResearch/hermes-agent/actions/runs/33187685770): all `action_required`. |

The head CI run has zero jobs. `action_required` is an authorization gate, **not a test failure and not a pass**. These receipts do not establish all-commit green. The canonical `contributors/emails/kbj2043@gmail.com` mapping requested earlier is still absent at the reviewed head and inspected main; preserve `Lukas <kbj2043@gmail.com>` attribution and add the mapping rather than rewriting authorship.

### Updated composition graph

DavidMetcalfe's [#96709](https://github.com/NousResearch/hermes-agent/pull/96709) remains open and overlaps the memory/prompt seams; its personal-memory filtering must not filter machine-wide GLOBAL policy. astraltrekkin's [#97061](https://github.com/NousResearch/hermes-agent/pull/97061) and 686f6c61's [#96500](https://github.com/NousResearch/hermes-agent/pull/96500) remain open on delegation lifecycle seams, alongside the open delegation-fracture issue [#79928](https://github.com/NousResearch/hermes-agent/issues/79928). Reconcile those shared files during the restack rather than duplicating owners. Correction to the earlier graph: omxhub's [#96255](https://github.com/NousResearch/hermes-agent/pull/96255) closed **unmerged on September 1**; it is no longer an open landing dependency.

The disabled-memory correction should stand. Session restoration still needs to join the same snapshot contract before the original coherence blocker can be closed.

agent/agent_init.py
```diff
@@ -1847,6 +1848,19 @@ def init_agent(
...
+    # Machine-wide policy is independent of the writable personal-memory
+    # stores. Snapshot it for every agent, including skip_memory/minimal agents,
+    # so policy application is not accidentally controlled by memory settings.
+    agent._global_policy_snapshot = global_policy_snapshot
+    if agent._global_policy_snapshot is None:
+        agent._global_policy_snapshot = ""
+        try:
+            from tools.memory_tool import load_global_policy_block
+
+            agent._global_policy_snapshot = load_global_policy_block()
```
andrexibiza: [P1] Restore the logical session's GLOBAL snapshot before delegating

This disk read happens for every new AIAgent, including reconstruction of an existing gateway session. Suppose session S persisted a system prompt containing policy A, then GLOBAL.md changes to B. The reconstructed instance sets `_global_policy_snapshot` to B here, but the ordinary unchanged-runtime continuation in [`_restore_or_build_system_prompt`](https://github.com/NousResearch/hermes-agent/blob/c69ecaa314e14282529a1df33c9b4bf3b5837d08/agent/conversation_loop.py#L998-L1021) restores S's full cached prompt containing A without restoring this new field. The [turn prologue](https://github.com/NousResearch/hermes-agent/blob/c69ecaa314e14282529a1df33c9b4bf3b5837d08/agent/turn_context.py#L820-L824) uses that restored prompt, while [`_build_child_agent`](https://github.com/NousResearch/hermes-agent/blob/c69ecaa314e14282529a1df33c9b4bf3b5837d08/tools/delegate_tool.py#L1744-L1761) forwards the field containing B to the child constructor.

The parent therefore still operates under A while its new child operates under B in the same logical session. A→missing is worse: the parent keeps A and the child gets an explicitly empty snapshot. The same-live-parent tests do not cover this reconstruction boundary.

Please persist/recover the session-owned snapshot together with the stored prompt, including an explicit empty snapshot, before delegation or rebuilding can consume it. A regression should reconstruct a real AIAgent for the same SessionDB session after changing/deleting/creating GLOBAL.md, then exercise the production child builder and assert parent/child policy equality. A genuinely new session must still pick up the new disk policy. Do not solve this by rereading policy only for children or by treating an empty frozen snapshot as uninitialized.

PR #97234, opened 2026-08-28
fix: make delegation retries idempotent and isolate MCP breaker

ugoi:
## Summary
- recover the original child handle when a launch acknowledgement is lost and the same correlation ID is retried
- serialize overlapping admission, reject changed requests, and clean up failed submissions
- classify MCP reachability from transport/RPC completion rather than rendered tool/application errors
- preserve breaker charging for EOF, timeout, connectivity, dead-stdio, disconnected-session, and backend exceptions

## Verification
- delegation baseline: 5 passed, 4 failed; fixed: 9 passed
- delegation relevant regression suite: 136 passed
- MCP breaker RED: rendered error incremented 0→1
- MCP breaker focused: 29 passed
- MCP tools suite: 599 passed
- Ruff and git diff --check passed

### DIFF
--- a/agent/subagent_lifecycle.py
+++ b/agent/subagent_lifecycle.py
@@ -145,6 +145,7 @@ class _Record:
     started_at: Optional[float] = None
     completed_at: Optional[float] = None
     result: Optional[SubagentResult] = None
+    request_fingerprint: Optional[str] = None
 
 
 class _Registry:
@@ -152,6 +153,7 @@ class _Registry:
 
     def __init__(self) -> None:
         self.lock = threading.RLock()
+        self.launch_lock = threading.RLock()
         self.records: dict[str, _Record] = {}
         self.correlations: dict[tuple[Optional[str], str], str] = {}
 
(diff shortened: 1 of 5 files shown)


Enough1122:

**Overall:** Makes delegation retries idempotent and isolates MCP breaker per launch.

**What it does**
- `agent/subagent_lifecycle.py` adds `launch_lock` (re-entrant), `request_fingerprint`, `correlations` map. `launch` serializes construction via `launch_lock` so two overlapping retries with same `correlation_id` don't build separate children; identical fingerprint returns existing handle, differing fingerprint raises `correlation_id was already used for a different request.` Also repairs dangling index entries.
- MCP breaker isolation so pre-existing breaker from parent doesn't leak into child — verified via sibling diffs; main registry change here is correlation correctness.

**Non-blocking notes**
- `request_fingerprint` derivation not visible in truncated diff — ensure it hashes goal/context/toolsets conservatively; otherwise identical correlation with slightly different context would incorrectly reuse.
- `launch_lock` is process-wide; status/cancel/result still use narrow `_REGISTRY.lock` — good to keep reads concurrent.
- Repair of dangling `correlations` entry (`pop` when `records.get(existing_id)` is None) prevents ghost id block.

Important retry correctness fix.

PR #2, opened 2025-08-31
Support passing morph snapshot id

hjc-puro:
Allows starting from different snapshot (creating new snapshot at the end is still todo)

```sh
 python run_agent.py -q "how much disk space do i have in the terminal tool vm" --morph_snapshot_id snapshot_h3jg94cd
```

How to make a new snapshot to test this:
```sh
hecate "clone this repo and show me the file tree git@github.com:axolotl-ai-cloud/axolotl.git" --save_snapshot_if_success --snapshot_id snapshot_p5294qxt
```
image
This command starts an instance from `snapshot_p5294qxt` and then clones axolotl into it. we then snapshot the instance and get a new task id.

If you would like to ssh into it and manually make changes and then snapshot again you can do this:
```sh
morphcloud instance start snapshot_p5294qxt
# returns instance id like morphvm_1234
morphcloud instance ssh morphvm_1234
# do stuff
morphcloud instance snapshot morphvm_1234
# returns snapshot id like snapshot_abcd
morphcloud instance stop morphvm_1234
```

PR #48852, opened 2026-06-19
fix(gateway): resolve watchdog, env override, and proxy routing on Windows

JoaoMarcos44:
### Overview & Context
This PR resolves three durability and configuration fragility bugs on Windows (Issue #48820).

This is a joint effort:
- Salvaged from the initial work on #48852 by `@JoaoMarcos44`.
- Co-authored and hardened by `@kshitijk4poor` in their salvage PR #48937 to resolve critical regressions in the initial watchdog and proxy monkeypatching.
- Further optimized by `@JoaoMarcos44` to reduce the number of modified files (which was too large reducing it from 19 to 13 files) by streamlining the explicit proxy isolation and cleaning up the commits.

#### Bug 1: Watchdog/Supervision of the Worker (Windows Daemon Loop)
* **Fix**: Added a Python-native watchdog process supervisor (`run_watchdog` loop in [gateway_windows.py](file:///c:/Users/Nitro/hermes-agent/hermes_cli/gateway_windows.py)).
  * Uses Win32 Job Object containment (`ctypes`) to automatically kill the child worker process and all descendant/browser helper processes when the watchdog exits.
  * Implements capped exponential backoff (1s -> 60s max) resetting on healthy runs (>=30s).
  * Implements a circuit breaker (max 5 retries in a rolling 60s window) to prevent infinite CPU-looping on bad config.
  * Implements automatic log rotation for `gateway-stdio.log` to `.old` if it exceeds 5MB, executing before the child process is spawned to avoid Windows file locks.
  * Handles exit code 0 as clean stop and exit code 75 as immediate worker restart.

#### Bug 2: `enabled: false` Ignored When Environment Variables Are Present
* **Fix**: Generalized the explicit config desativation marker (`_enabled_explicit`) across all platforms (built-ins and plugins) during config load, and updated it to use `.get()` instead of `.pop()` in [config.py](file:///c:/Users/Nitro/hermes-agent/gateway/config.py). This ensures subsequent registry loops do not override the user's explicit disablement intent.

#### Bug 3: `trust_env=True` Routing Traffic to System Proxy in Scheduled Tasks
[...]

Issue #48820, opened 2026-06-19
[Windows] Messaging gateway has 3 fragility bugs: no watchdog, env override ignores yaml, aiohttp trust_env misroutes under Scheduled Task


ranlingfeng:
# Hermes messaging gateway on Windows is fragile: three related issues causing silent failure

## Environment

- OS: Windows 10 / Windows 11
- Hermes install path: `%LOCALAPPDATA%\hermes` (git-installed from `NousResearch/hermes-agent`)
- Hermes commit: `b39ec2fc3` on `main` (2026-06-18), with 47 upstream commits behind
- Python: 3.11.15 (uv-managed cpython inside the venv)
- Local proxy present: Clash / FlClashCore / V2RayN listening on `127.0.0.1:7890`
  (TUN mode or system proxy)

## Summary

Three related bugs make the Hermes **messaging gateway** (the part that polls
Telegram / Discord / Weixin / etc., distinct from the desktop GUI dashboard)
unreliable on Windows. They were observed together in a single session, but
each is independently fixable.

---

### Process tree

The gateway is launched by Windows Scheduled Task `Hermes_Gateway` via
`%LOCALAPPDATA%\hermes\gateway-service\Hermes_Gateway.cmd`:

```
Task Scheduler (Logon Trigger, fires ONCE at user login)
  └─ cmd.exe (PID A, runs Hermes_Gateway.cmd)
       └─ pythonw.exe (PID B, `-m hermes_cli.main gateway run`)
            └─ uv-managed cpython (PID C, the actual worker)
```

### Trigger configuration

```powershell
PS> Get-ScheduledTask -TaskName 'Hermes_Gateway' | Format-List Triggers, Settings
Triggers[0] : MSFT_TaskLogonTrigger   # fires once at login, no repetition
Settings.MultipleInstances : IgnoreNew
Settings.ExecutionTimeLimit : PT72H     # 72h hard kill, not the cause here
```

### Failure signature

When PID C receives SIGINT (from a `hermes gateway restart`, a GUI healthcheck,
or any other source):

1. PID C exits
2. PID B (the `pythonw.exe` wrapper) **does not restart** PID C; it exits silently
3. The Scheduled Task sees its process tree end and transitions to `Ready` state
4. **No automatic restart occurs** — `Logon Trigger` only fires once, and the
   wrapper does not supervise the worker

Observables:

- `hermes gateway status` → `✗ No gateway process detected`
[...]

PR #97237, opened 2026-08-28
fix(update): verify Windows gateway relaunch and keep a stderr trace (#48820)

tutan0558:
## Summary

The Windows update pause/resume path trusted the spawn call: a respawned gateway could die within seconds while the update printed `✓ Restarting Windows gateway profile(s)` — and nothing anywhere recorded the death. Reproduced **4 times** on a real install (#48820): no gateway.log line, no exit-diag record, no WER, no Defender detection, because the watcher respawns with `stdout=DEVNULL, stderr=DEVNULL`.

This PR is a focused fix for the silent-death class, on current main:

1. **Verify the relaunch instead of trusting it.** `_resume_windows_gateways_after_update` now polls `gateway_state.json` for a fresh pid (not one of the pre-update pids) that is actually alive in the process table (25s budget). If nothing comes up, the update prints an explicit ⚠ warning with the manual relaunch command (`wscript <hermes_home>\gateway-service\Hermes_Gateway.vbs`) instead of a misleading ✓.

2. **Keep a stderr trace for respawned gateways.** The watcher's respawn now writes stderr to `hermes/logs/gateway-respawn-stderr.log` instead of DEVNULL, so a gateway that dies on startup leaves evidence for the next diagnosis.

## Why this instead of salvaging #48852

#48852 (open, "salvageability: medium" per the sweeper) is a larger watchdog/env/proxy rework based on June's main; its patch no longer applies to current main (files moved/rewritten: `gateway/platforms/matrix.py`, `slack.py`, `config.py`, `run.py` etc. — `git apply` fails on 8 files). This PR deliberately does not touch the watchdog architecture; it closes the worst UX gap of the reported class (silent success + silent death) with a small, reviewable change.

## Reproduction context (for the record)

2026-08-28, v0.20.5 → v0.20.6 hand-off update: old gateway (pid 41672) stalled during shutdown (weixin mid backoff-retry, bad network), was force-killed (`lifecycle_ledger`: "exited UNCLEANLY (no exit path ran — SIGKILL)"). Updater respawned pid 48452 at 08:40:06; it wrote its `gateway.start` exit-diag entry and died within seconds — zero trace. Control: same venv, same machine, gateway launched via the vbs script stays up for hours. Strong suspect is job-object teardown when the updater process exits (the watcher comment at `gateway.py` warns about exactly this: without `CREATE_BREAKAWAY_FROM_JOB` the respawned gateway dies when the parent job tears down), but no crash dump exists to prove it — which is why fix #2 (stderr trace) matters.

## Testing

- `python -m py_compile` on both modified files: OK
- Functional smoke test of `_verify_windows_gateway_relaunch` against the live machine: with `previous_pids={41672, 48452}` it correctly detected the currently-running gateway (pid 65240) as alive and returned `True`

### DIFF
--- a/hermes_cli/gateway.py
+++ b/hermes_cli/gateway.py
@@ -1348,9 +1348,16 @@ def _spawn_gateway_restart_watcher(old_pid: int, run_argv: list[str]) -> bool:
         # been spawned inside a job object (Electron/Tauri parent), and
         # without breakaway the respawned gateway would die when that job
         # tears down. See _subprocess_compat.windows_detach_flags().
+        # Keep a stderr trace for the respawned gateway: with both streams
+        # in DEVNULL, a gateway that dies on startup leaves no evidence at
+        # all (#48820 — reproduced 4x, zero trace each time).
+        _respawn_stderr = os.path.join(
+            os.environ.get("LOCALAPPDATA") or os.path.expanduser("~"),
+            "hermes", "logs", "gateway-respawn-stderr.log",
+        )
         _popen_kwargs = {{
             "stdout": subprocess.DEVNULL,
-            "stderr": subprocess.DEVNULL,
+            "stderr": open(_respawn_stderr, "ab"),
         }}
         # Anchor the respawned gateway at the stable working dir and overlay
         # the env (VIRTUAL_ENV / PYTHONPATH / HERMES_HOME) the windowless
--- a/hermes_cli/update_cmd.py
+++ b/hermes_cli/update_cmd.py
@@ -6825,6 +6825,38 @@ def _refresh_bootstrap_cache_scripts(branch: str = "main") -> None:
     except Exception as exc:
         logger.debug("Could not refresh bootstrap-cache scripts after update: %s", exc)
 
+def _verify_windows_gateway_relaunch(previous_pids, timeout_s=25):
+    """Wait for a fresh gateway pid to appear in gateway_state.json and be alive.
+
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Surfaces silent Windows gateway respawn failures (#48820) and preserves stderr.

**What it does**
- `hermes_cli/gateway.py:_spawn_gateway_restart_watcher` now opens `LOCALAPPDATA/hermes/logs/gateway-respawn-stderr.log` in `ab` instead of `DEVNULL` for `stderr`; stdout stays `DEVNULL`. Keeps trace when respawned gateway dies on startup.
- `hermes_cli/update_cmd.py:_verify_windows_gateway_relaunch` polls `gateway_state.json` up to 25s for a `pid not in previous_pids` with `_pid_exists(pid)` true; returns `(True,pid)` or `(False,None)` to turn silent death into warning instead of "✓ Restarting".

**Non-blocking notes**
- `open(_respawn_stderr, "ab")` without explicit close — handle inherited by child process; parent fd remains open. Consider `open(..., "ab", buffering=0)` or closing in parent after `Popen` to avoid fd leak.
- Poll interval 2s up to 25s — matches typical Windows SCM start latency; good.
- `get_hermes_home()` respects profile overlay — polling correct profile's `gateway_state.json`.

Good observability fix for flaky respawn.

teknium1:
Windows gateway watchdog relaunch is fixed on main by ab9866bc64 (#100594, proven on windows-latest). Thanks for the repro, @tutan0558; closing as superseded.

PR #91104, opened 2026-08-20
fix(agent): replay plaintext Responses reasoning for relays without encrypted continuation blobs

Tom-A-Lynch:
## What does this PR do?

Fixes a silent reasoning-history loss that degenerates agents into tool-call loops when Hermes runs in Responses mode against a relay that returns **plaintext** reasoning instead of OpenAI's `encrypted_content` continuation blob.

The failure chain:

1. A Responses relay (e.g. Actual Computer's `api.actual.inc`, serving Kimi-K3 via vLLM) cannot mint OpenAI's opaque encrypted reasoning blob, so it returns model thinking as plaintext `reasoning` content.
2. Hermes's Responses adapter *displays* that reasoning but only *replays* reasoning items carrying `encrypted_content`. Plaintext reasoning is silently dropped from the next request.
3. The backend then renders historical assistant turns as an empty think block plus the tool call (`<think></think>` + action). Models whose chat templates attach thinking to the action turn (Kimi-K3) see malformed action-only history and collapse into repeated tool calls / infinitely growing greps.

The fix: when the route targets a plaintext-reasoning relay, capture the plaintext reasoning from the response and replay it as a `reasoning` item **at the head of the assistant turn it belongs to**, in OpenAI's canonical output order (`reasoning` → assistant message, if any → `function_call`(s)), surviving both preflight passes. The adjacency contract is turn-level: the relay attaches the reasoning to the contiguous assistant run that follows it (the whole model turn), so mixed text+tool-call turns and parallel tool calls replay losslessly too — pinned by regression tests, including the mixed-content shape raised in review.

Safety properties (all pinned by tests):

- Encrypted reasoning always takes precedence; the same thought is never replayed twice.
- The capability is gated to `provider=actual` or the exact `api.actual.inc` hostname (same idiom as the existing Azure Foundry hostname gate). Lookalike hosts (`api.actual.inc.evil.example`, path-embedded) do not qualify.
- Native OpenAI routes are byte-for-byte unaffected; the capability resets on provider switch within a session.
- Reasoning items keep a following item, so `missing_following_item` cannot trigger — and the empty-assistant-message guard never injects a synthetic message into a turn that has a `function_call`, which would split the run.

## Related Issue

No existing issue; searched open/closed PRs and issues for duplicates and found none. Happy to file one if preferred.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

[...]

PR #97241, opened 2026-08-28
fix(responses): replay DeepSeek plaintext reasoning

franckrocca:
## Summary
- classify `api.deepseek.com` as a stable Responses issuer
- persist DeepSeek plaintext `reasoning_text` items when `encrypted_content` is null
- replay the provider-verified minimal `type + content` shape after tool calls
- drop those items across issuer changes

## Root cause
DeepSeek V4's stateless Responses API requires plaintext `reasoning_text` to be passed back after tool calls. Hermes previously persisted and replayed only encrypted reasoning carriers, so the second tool turn failed with HTTP 400.

## Tests
- `76 passed` — DeepSeek regression + adjacent Codex/xAI/provider parity
- `279 passed` — `tests/run_agent/test_run_agent.py`
- `compileall`, `git diff --check`, and Gitleaks pass

One pre-existing `_BarrierDB.flush_token_counts` thread warning remains in `test_run_agent.py`; no test failed.

### DIFF
--- a/agent/codex_responses_adapter.py
+++ b/agent/codex_responses_adapter.py
@@ -18,6 +18,7 @@ import unicodedata
 import uuid
 from types import SimpleNamespace
 from typing import Any, Dict, List, NamedTuple, Optional
+from urllib.parse import urlparse
 
 from agent.message_sanitization import deterministic_call_id
 from agent.prompt_builder import DEFAULT_AGENT_IDENTITY
@@ -48,7 +49,10 @@ def _classify_responses_issuer(
     if is_codex_backend:
         return "codex_backend"
     if base_url:
-        return f"other:{base_url}"
+        hostname = (urlparse(base_url).hostname or "").lower()
+        if hostname == "api.deepseek.com":
+            return "deepseek_responses"
+        return f"other:{base_url.rstrip('/')}"
     return "other"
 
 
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Replays DeepSeek plaintext reasoning through Responses API.

**What it does**
- `agent/codex_responses_adapter.py:_classify_responses_issuer` now normalizes `base_url` via `urlparse(hostname)`; `api.deepseek.com` classifies as `deepseek_responses` vs generic `other:{base_url.rstrip('/')}`.
- `_chat_messages_to_responses_input` treats reasoning items without `encrypted_content` as replayable when `_issuer_kind=="deepseek_responses"` and `_normalize_plaintext_reasoning_content(ri.get("content"))` truthy. Replay shape is minimal `{"type":"reasoning","content":…}` stripping `id/_issuer_kind/status/summary` since DeepSeek stateless surface requires `reasoning_text` after every tool call.

**Non-blocking notes**
- `urlparse(hostname)` lowercases host — good for mixed-case `API.DeepSeek.COM`.
- `encrypted_reasoning = bool(ri.get("encrypted_content"))` — if DeepSeek ever sends `encrypted_content: null` vs missing, both false and plaintext path triggers; correct.
- Ensure `replay_item` for encrypted+plaintext hybrid keeps encrypted form (code does `plaintext and not encrypted` guard).

Correct provider-specific replay fix.

Malvix:
Cross-reference: same bug class observed on a third gateway (Aperture / Vercel AI Gateway shape serving `moonshotai/kimi-k3`): plaintext `reasoning_text` items with `encrypted_content` null are never persisted or replayed, so the thinking model loses its chain after the first tool call. #91104 covers `api.actual.inc`, this PR covers `api.deepseek.com`.

I opened #100918 with the host-agnostic form: capture and replay plaintext reasoning for any issuer that is not `codex_backend` / `xai_responses` / `github_responses`, so a new gateway does not need its own issuer entry. It mirrors the received `summary`/`content` shape rather than a minimal `type + content`, and stamps `_issuer_kind` the same way this PR does so the cross-issuer guard still drops items on a provider switch. The DeepSeek round-trip test here would be a good addition to that suite if maintainers prefer to consolidate; happy to go either way.

PR #97243, opened 2026-08-28
fix(curator): prevent overlapping review passes

chandlerm923:
## What does this PR do?

Prevents multiple curator review passes from running concurrently for the same Hermes profile.

Scheduled and manual curator runs could overlap before curator state was updated, causing duplicate LLM work and concurrent skill mutations. This change adds a profile-scoped, non-blocking file lock that remains held for the entire review pass, including background execution.

A concurrent invocation now exits cleanly without entering the review.

## Type of Change

- [x] Bug fix

## Changes Made

- Add cross-process curator locking using `fcntl` on POSIX and `msvcrt` on Windows
- Hold the lock until the background review thread finishes
- Release the lock on both successful and exceptional exits
- Report concurrent manual runs clearly in the CLI
- Add unit, CLI, and real two-process E2E coverage

No dependencies, configuration keys, or environment variables were added.

## How to Test

```bash
python -m pytest tests/agent/test_curator.py tests/agent/test_curator_backup.py tests/hermes_cli/test_curator_run.py tests/e2e/test_curator_single_flight.py -q
```

### DIFF
--- a/agent/curator.py
+++ b/agent/curator.py
@@ -21,6 +21,7 @@ Strict invariants:
 
 from __future__ import annotations
 
+import errno
 import json
 import logging
 import os
@@ -35,6 +36,16 @@ from tools import skill_usage
 from utils import atomic_json_write
 
 logger = logging.getLogger(__name__)
+
+try:  # pragma: no cover - platform-dependent import
+    import fcntl
+except ImportError:  # pragma: no cover - Windows
+    fcntl = None  # type: ignore[assignment]
+
+try:  # pragma: no cover - platform-dependent import
+    import msvcrt
+except ImportError:  # pragma: no cover - POSIX
+    msvcrt = None  # type: ignore[assignment]
 
 
 def _strip_aux_credential(value: Any) -> Optional[str]:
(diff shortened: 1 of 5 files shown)


Enough1122:

**Overall:** Serializes curator passes with cross-process file lock.

**What it does**
- `agent/curator.py:_CuratorRunLock` wraps `fcntl.flock` (POSIX) / `msvcrt.locking` (Windows) over `get_hermes_home() / ".curator-run.lock"` (outside skills tree so rollback doesn't wipe it). `_try_acquire_curator_run_lock` opens `a+b`, ensures byte 0 exists, seeks to 0 before lock, raises `RuntimeError` if neither backend available, treats `EACCES/EAGAIN/EWOULDBLOCK/EDEADLK` as contention (`None`) else propagates.
- Curator entry points try_acquire at pass start and release in `finally`; overlapping `hermes curator run` / cron triggers now skip or wait instead of interleaving LLM reviews and archive moves.

**Non-blocking notes**
- Lock file is `a+b` — creation with `parents mkdir` good; ensure `msvcrt` byte-0 seek before both lock and unlock (code does seek before lock; release also seeks to 0).
- `_is_run_lock_contention` checks `errno is None` -> False — correctly doesn't misclassify `errno` missing as contention.
- `handle.close()` in `release` finally — fd closed once; safe for re-entry after `self._handle = None`.

Prevents overlapping review corruption; correct platform branching.

PR #97244, opened 2026-08-28
feat(desktop): support multi-path and directory pickers in the plugin OS API

charliechen1113-cpu:
## Summary

- extend the existing `ctx.os.pickOpenPath(...)` options with `directories` so plugins can ask the native chooser for one folder
- add `ctx.os.pickOpenPaths(...)` for user-mediated selection of multiple files or directories
- preserve the current result-shaped picker convention: single-path pickers return `null`, while the multi-path picker returns `[]`, on cancellation, failure, or an unavailable older shell
- export the file-dialog option types through `@hermes/plugin-sdk`
- document the complete plugin path-picker surface, including the existing open and save helpers

## Motivation

The existing `pickOpenPath` and `pickSavePath` helpers cover one input file and one save destination. Desktop plugins also need the same curated OS boundary for workflows that accept several source files or a user-selected directory.

This update extends the current `PluginOs` surface rather than exposing `window.hermesDesktop` or adding a second overlapping bridge API.

## API

```ts
ctx.os.pickOpenPath({
  title?,
  defaultPath?,
  directories?,
  filters?
}) // Promise<string | null>

ctx.os.pickOpenPaths({
  title?,
  defaultPath?,
  directories?,
  filters?
}) // Promise<string[]>
```

`pickOpenPaths` always requests native multiple selection. Setting `directories: true` on either open picker selects folders instead of files.

## Security and compatibility

- plugins remain behind the curated `PluginContext` boundary and receive no raw access to `window.hermesDesktop`
- the chooser only returns paths explicitly selected by the user
- no directory scanning or arbitrary filesystem read capability is added
- older Desktop shells continue to degrade safely through result values instead of throwing

## Tests

- missing Desktop bridge returns an empty list for `pickOpenPaths`
- multiple selected paths are returned and `multiple: true` is passed to the native bridge
- `directories: true` is passed through for single-directory selection
- native bridge failure resolves to an empty list
- targeted plugin OS tests pass
- Desktop TypeScript checks pass
- Desktop ESLint passes with no errors

### DIFF
--- a/apps/desktop/src/contrib/plugin.test.ts
+++ b/apps/desktop/src/contrib/plugin.test.ts
@@ -41,6 +41,7 @@ describe('createPluginContext.os', () => {
     // The pickers answer with a path, so their "unavailable" is null.
     await expect(ctx.os.pickSavePath()).resolves.toBeNull()
     await expect(ctx.os.pickOpenPath()).resolves.toBeNull()
+    await expect(ctx.os.pickOpenPaths()).resolves.toEqual([])
   })
 
   it('file pickers return the chosen path, and null on cancel', async () => {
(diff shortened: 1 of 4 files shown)


Enough1122:

**Overall:** Plugin OS API now exposes native file/directory picker.

**What it does**
- `apps/desktop/src/contrib/plugin.ts:PluginOs` adds `selectPaths(options?: HermesSelectPathsOptions) => Promise<string[]>`; implementation routes through `window.hermesDesktop.bridge.selectPaths` when available, else `[]`. Contract: cancellation / unavailable older shell / malformed output → `[]`; real bridge throw → rejected promise surfaced to plugin.
- Tests `plugin.test.ts` pin `[]` when no bridge, routing + success + failure propagation.

**Non-blocking notes**
- `selectPaths` returning `[]` on unavailable vs throwing on failure is a deliberate split — good to document for plugin authors distinguishing user cancel vs error.
- Type import `HermesSelectPathsOptions` from `@/global` — ensure global augments `HermesDesktop` with `selectPaths` to keep TS strict.

Clean plugin surface extension.

PR #92325, opened 2026-08-22
fix(codex): keep sessions usable after provider switches

fangliquanflq:
## What does this PR do?

Prevents a live session from becoming unusable on the ChatGPT Codex Responses backend after switching from another provider. Replayed assistant message items now omit foreign short IDs that Codex rejects, while native `msg*` IDs remain available for prefix-cache hits and other Responses issuers keep their existing behavior.

### Symptom

After a provider switch, Codex rejects every subsequent turn with HTTP 400: `Invalid input[N].id ... Expected an ID that begins with 'msg'`.

### Impact

Sessions containing a foreign assistant message ID cannot continue on Codex until the user starts a new session. The persisted invalid item is replayed on every later turn.

### Bug Cause

**Trigger:** `agent/codex_responses_adapter.py` in `_chat_messages_to_responses_input`, when replaying `codex_message_items` for `current_issuer_kind == "codex_backend"`.

**Causal chain:**
1. A non-Codex provider stores a short assistant message ID such as a UUID.
2. The replay converter accepts it because it is non-empty and no longer than 64 characters.
3. The ChatGPT Codex backend requires message IDs to begin with `msg`, rejects the request with HTTP 400, and the persisted ID breaks later turns too.

**Why it is wrong:** The existing replay guard validates only length and the Copilot connection policy, not the stricter Codex message-ID namespace.

**Working sibling / contrast:** Native Codex `msg*` IDs already satisfy the backend contract and remain useful for prefix caching. xAI and custom Responses issuers do not use this Codex-only prefix rule.

**Ruled out:** This is not the oversized-ID replay bug: a 36-character UUID is below the existing 64-character limit and still reproduces the invalid-value path.

### Fix

Apply the `msg` prefix requirement only when the current issuer is the ChatGPT Codex backend. Preserve valid native IDs, omit foreign IDs without altering message content, phase, or status, and leave other issuer policies unchanged.

## Related Issue

Closes #92311

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `agent/codex_responses_adapter.py` - filter foreign assistant message IDs during Codex replay.
- `tests/agent/test_codex_responses_adapter.py` - cover Codex filtering and xAI preservation at the converter boundary.
- `tests/agent/transports/test_codex_transport.py` - cover the complete transport kwargs path.

## How to Test

1. Run the adapter regression suite.
2. Run the Codex transport regression suite.

```bash
[...]

PR #97246, opened 2026-08-28
fix(agent): guard Codex message IDs across providers

### LINKED ISSUE #92311: [Bug]: Codex Responses 400 when replaying foreign input[].id that does not start with msg after mid-session provider switch
## Bug Description

Switching a live session from a non-OpenAI Responses provider (for example xAI Grok or a Chat Completions model) to `openai-codex` (GPT-5.x Luna / Codex) fails with a non-retryable HTTP 400.

The Codex backend requires assistant `input[].id` values to begin with `msg`. Hermes replays stored `codex_message_items[].id` as-is when the string is ≤ 64 characters. Foreign issuers often mint short UUIDs (36 chars). Those pass the existing length guard and then get rejected by Codex, which bricks the session for Luna/Codex until `/new`.

This is the same family as the existing `input[].id` replay bugs, but a different validation rule:

- #27038 / #10788 / #17318 — drop IDs longer than 64 characters
- #32716 — drop Copilot connection-bound IDs
- #32617 / #34205 — drop cross-issuer `encrypted_content` on provider switch

None of those guards check the `msg` prefix. Function-call IDs already get rewritten to start with `fc_` (`_derive_responses_function_call_id`). Message IDs do not get the equivalent treatment.

## Steps to Reproduce

1. Start a gateway session on a non-OpenAI Responses provider (xAI Grok, or a Chat Completions model).
2. Have a multi-turn conversation so the session stores assistant `codex_message_items` whose `id` is a UUID (or any id that does not start with `msg`).
3. Mid-session, switch the model to `openai-codex` / `gpt-5.6-luna` (or any Codex Responses model).
4. Send another message in the same session.

## Expected Behavior

Hermes should make the replayed payload valid for the *current* Responses issuer. For official Codex this means:

- omit `id` on replayed assistant message items that are not Codex-minted, or
- rewrite them to a `msg_…` surrogate (same idea as `_derive_responses_function_call_id` for `fc_`)

A mid-session provider switch should not permanently brick the session.

The same session should still work on xAI / Chat Completions, which do not enforce this prefix.

## Actual Behavior

Codex rejects the request. Hermes classifies it as a non-retryable client error. Every later Luna/Codex turn in that session fails the same way. Fallback to Grok or a Chat Completions model still works because those backends either ignore the id or strip the sidecar.

## Environment

- OS: Linux (Debian)
- Hermes: v0.20.5 (2026.8.19), git install
- Python: 3.11
- Providers involved: xAI OAuth / Chat Completions → `openai-codex` (`chatgpt.com/backend-api/codex`)

## Error Output

```
HTTP 400: Invalid 'input[N].id': '<uuid>'. Expected an ID that begins with 'msg'.
error.type=invalid_request_error
error.param=input[N].id
error.code=invalid_value
```

Agent log classifies this as `BadRequestError` / non-retryable. No retry, no fallback sanitizer.

## Additional Context

In `agent/codex_responses_adapter.py`, replay keeps `id` when:

- not GitHub Copilot, and
- `id` is a non-empty string, and
- `len(id) <= _MAX_RESPONSES_ITEM_ID_LENGTH` (64)

There is no `startswith("msg")` check. Comments even say short Hermes-minted `msg_...` ids are kept on purpose for prefix-cache hits. That is correct for same-issuer Codex ids, and wrong for foreign short ids.

Suggested fix (smallest): when `current_issuer_kind` is official Codex/OpenAI, drop or rewrite any replayed message `id` that does not start with `msg`. Leave same-issuer `msg_…` ids intact so prefix cache still works.

I did not find an existing issue for this exact `invalid_value` / `begins with 'msg'` error.

Amekn:
## What does this PR do?

Prevents a session from becoming unusable when switching from another Responses-compatible provider to OpenAI Codex. A foreign assistant message ID can satisfy the existing length limit but still cause `HTTP 400: Expected an ID that begins with msg`.

Persisted assistant message items now carry issuer provenance. Replay withholds foreign IDs while preserving assistant text, phase and status. Unstamped legacy history remains compatible, except that Codex only receives IDs in its required `msg` namespace. Custom endpoint URLs are canonicalized consistently with the OpenAI SDK so the main and auxiliary paths agree.

## Related Issue

Fixes #92311.

Related to #92325. This PR additionally protects cross-issuer replay through persisted provenance, auxiliary/custom-proxy routes, and SDK URL normalization.

## Type of Change

## Changes Made

Rebased the original commit onto `67764dc0863349a384c16425e73ee8571f3a94b7`, preserving its author and author date. Current head: `f538ae315c126e7dd07876970d7a77834bb5cd88`.

- `agent/codex_responses_adapter.py`: port the fix to the current replay helpers and `_OutputScan`; stamp persisted assistant items; apply issuer and Codex namespace checks; canonicalize legacy URL-backed message/reasoning stamps without mutating stored history or forwarding internal metadata.
- `agent/auxiliary_client.py`: carry explicit Codex/xAI issuer identity through the current auxiliary request builder, including custom proxies; apply the same shared replay policy as the main transport.
- `tests/agent/test_responses_issuer_replay.py`: consolidate regression coverage into two parameterized behavior tests, following the current contribution guidance. Covers legacy and stamped histories, cross-provider replay, ID length and namespace, content/phase/status preservation, URL equivalence and malformed-URL fallback, and real SDK HTTP/SSE exchanges through the main and auxiliary paths.

Provider classification remains flag-based. URL normalization deliberately does not promote a custom `other:` URL to a named provider category. The Codex check deliberately uses `startswith("msg")`, not `startswith("msg_")`.

## How to Test

```bash
scripts/run_tests.sh \
  tests/agent/test_responses_issuer_replay.py \
  -j 1 --file-retries 0

scripts/run_tests.sh \
  tests/agent/test_codex_responses_adapter.py \
  tests/agent/test_auxiliary_client.py \
  tests/agent/transports/test_codex_transport.py \
  tests/run_agent/test_codex_xai_oauth_recovery.py \
  tests/run_agent/test_provider_parity.py \
  tests/run_agent/test_run_agent_codex_responses.py \
  tests/run_agent/test_native_compaction*.py \
  -j 4 --file-retries 0

ruff check agent/codex_responses_adapter.py \
  agent/auxiliary_client.py \
  tests/agent/test_responses_issuer_replay.py
```

### Verification for the rebased revision

| Check | Result |
| --- | --- |
| New regression cases on unmodified upstream base | 21 failed, 4 passed, 0 collection/runtime errors |
| Same regression cases with the fix | 25 passed, 0 failed |
| Related adapter, auxiliary, transport, provider and native-compaction suites | 602 passed, 0 failed |
| Total targeted tests | **627 passed, 0 failed**, with file retries disabled |
| Ruff on changed files | Passed |
| Git whitespace/conflict checks | Passed |

[Baseline reproduction run](https://github.com/Amekn/hermes-agent/actions/runs/34493333442) · [Rebased validation run](https://github.com/Amekn/hermes-agent/actions/runs/34493831901)

The wire tests use the real OpenAI SDK and a local HTTP/SSE server, with only credential sources replaced. They do not use live provider accounts. The canonical test runner provides isolated Hermes state and strips inherited credentials.

**The full repository suite was not rerun for this revision.** The targeted results above supersede the earlier submission's focused-test count; they are not a claim that every repository test has passed.

### Code

- [x] Read the current Contributing Guide and applicable AGENTS.md instructions
- [x] Conventional commit message; original contribution authorship preserved
- [x] Related issue and overlapping PR identified above
- [x] Only changes related to this fix; one commit and three changed files
- [x] Regression behavior demonstrated failing on the upstream base and passing with the fix
- [x] Relevant existing suites pass using `scripts/run_tests.sh`
- [ ] Full repository test suite passes on this rebased revision: not run
- [x] Tested on a GitHub-hosted Ubuntu 24.04 runner

- [x] Relevant docstrings updated; no user-facing configuration changes
- [x] No config example, architecture guide or tool schema updates required
- [x] Cross-platform impact considered; conversion is platform-neutral
- [x] Temporary validation workflows are not included in this PR

## Reproduction / Expected Result

Before: a foreign short assistant message ID is replayed to Codex and triggers a non-retryable HTTP 400.

After: the same history retains assistant output text, phase and status, but omits the invalid or foreign ID. Same-issuer valid IDs remain available for prefix-cache reuse, and foreign encrypted reasoning is not replayed.

### DIFF
--- a/agent/auxiliary_client.py
+++ b/agent/auxiliary_client.py
@@ -1312,9 +1312,10 @@ class _CodexStreamGuard:
 class _CodexCompletionsAdapter:
     """Drop-in shim routing chat.completions.create() kwargs through Codex Responses streaming."""
 
-    def __init__(self, real_client: OpenAI, model: str):
+    def __init__(self, real_client: OpenAI, model: str, *, issuer_kind: Optional[str] = None):
         self._client = real_client
         self._model = model
+        self._issuer_kind = issuer_kind
 
     def _build_responses_kwargs(self, kwargs: Dict[str, Any]) -> Tuple[Dict[str, Any], str, Any]:
         """chat.completions kwargs → Responses API kwargs, ``(resp_kwargs, model, timeout)``; mirrors codex.py::build_kwargs."""
@@ -1328,12 +1329,18 @@ class _CodexCompletionsAdapter:
         # includes assistant tool_calls + role="tool" results). The shared converter encodes assistant tool
         # calls as `function_call` items and tool results as `function_call_output` items with a valid
         # call_id, so every Responses path normalizes tool history identically and cannot drift.
-        from agent.codex_responses_adapter import _chat_messages_to_responses_input
+        from agent.codex_responses_adapter import _chat_messages_to_responses_input, _classify_responses_issuer
         model = kwargs.get("model", self._model)
         host = str(getattr(self._client, "base_url", "") or "")
         is_xai = base_url_host_matches(host, "x.ai") or base_url_host_matches(host, "api.x.ai")
         is_copilot = base_url_host_matches(host, "githubcopilot.com")
         is_github = is_copilot or base_url_host_matches(host, "models.github.ai")
+        # Explicit provider identity survives custom proxy URLs; generic clients
+        # derive it from the endpoint, just like the main Responses transport.
+        issuer_kind = self._issuer_kind or _classify_responses_issuer(
+            is_xai_responses=is_xai, is_github_responses=is_github,
+            is_codex_backend=_is_official_codex_base_url(host), base_url=host,
+        )
         # System → ``instructions``; the rest goes through the SINGLE shared chat→Responses
         # converter (a private loop here once let role="tool" leak into input[]; the shared one
         # encodes tool history as function_call/function_call_output).
@@ -1352,7 +1359,8 @@ class _CodexCompletionsAdapter:
         # instead of agent/transports/codex.py's build_kwargs, so they need the same guard applied
         # independently. See #32716.
         input_items = _chat_messages_to_responses_input(
-            replay_messages, is_github_responses=is_copilot, native_compaction_eligible=False
+            replay_messages, is_github_responses=is_github,
+            current_issuer_kind=issuer_kind, native_compaction_eligible=False,
         )
         resp_kwargs: Dict[str, Any] = {
             # Codex only knows the base slug; strip the Hermes ``-900k`` picker suffix.
@@ -1539,9 +1547,9 @@ _AsyncAnthropicCompletionsAdapter = _AsyncCompletionsAdapter  # imported by test
 class CodexAuxiliaryClient:
     """OpenAI-client-compatible wrapper routing through the Codex Responses API (.api_key/.base_url for introspection)."""
 
-    def __init__(self, real_client: OpenAI, model: str):
+    def __init__(self, real_client: OpenAI, model: str, *, issuer_kind: Optional[str] = None):
         self._real_client = real_client
-        self.chat = _ChatShim(_CodexCompletionsAdapter(real_client, model))
+        self.chat = _ChatShim(_CodexCompletionsAdapter(real_client, model, issuer_kind=issuer_kind))
         self.api_key = real_client.api_key
         self.base_url = real_client.base_url
 
@@ -2729,7 +2737,7 @@ def _build_xai_oauth_aux_client(model: str) -> Tuple[Optional[Any], Optional[str
     real_client = _create_openai_client(
         api_key=api_key, base_url=base_url, default_headers=hermes_xai_default_headers()
     )
-    return CodexAuxiliaryClient(real_client, model), model
+    return CodexAuxiliaryClient(real_client, model, issuer_kind="xai_responses"), model
 
 
 def _build_codex_client(model: str) -> Tuple[Optional[Any], Optional[str]]:
@@ -2757,7 +2765,7 @@ def _build_codex_client(model: str) -> Tuple[Optional[Any], Optional[str]]:
         api_key=codex_token, base_url=base_url,
         default_headers=_codex_cloudflare_headers(codex_token, base_url=base_url),
     )
-    return CodexAuxiliaryClient(real_client, model), model
+    return CodexAuxiliaryClient(real_client, model, issuer_kind="codex_backend"), model
 
 
 def _try_azure_foundry(
--- a/agent/codex_responses_adapter.py
+++ b/agent/codex_responses_adapter.py
@@ -12,6 +12,8 @@ import uuid
 from types import SimpleNamespace
 from typing import Any, Callable, Dict, Iterator, List, NamedTuple, Optional, TypeGuard
 
+import httpx
+
 from agent.message_sanitization import deterministic_call_id
 from agent.prompt_builder import DEFAULT_AGENT_IDENTITY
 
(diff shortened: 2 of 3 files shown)


alt-glitch:

Related to #92325 and its issue #92311. Both prevent foreign message IDs from reaching Codex; this PR additionally stamps issuer provenance and covers auxiliary/custom-endpoint paths.

Enough1122:

This is a careful fix with strong test coverage. The core insight is right: a persisted `codex_message_items`/reasoning blob minted by one endpoint, replayed against another, causes hard 400s (invalid_encrypted_content / rejected ids), so stamping provenance on items and gatekeeping by issuer at replay is the correct seam. The `httpx.URL` normalization (lowercase host, drop default ports, strip trailing slash) correctly makes the SDK-normalized `client.base_url` agree with the configured one, and the legacy-`other:` canonicalization gives pre-upgrade sessions a migration path instead of silently dropping their reasoning. The `msg`-prefix rule for `codex_backend` is the right fail-safe.

A few points:

1. **`_canonicalize_responses_issuer_kind` re-runs `_classify_responses_issuer` on every stamped item in the replay loop.** For each reasoning item and message item, the code canonicalizes both `current_issuer_kind` (hoisted once, good) and the item stamp (per-item). Known labels (`xai_responses`, `codex_backend`, `github_responses`) short-circuit, so only `other:` items pay an `httpx.URL` parse — which is rare. Minor, but for a very long custom-endpoint session this is O(n) URL parsing every turn; a tiny per-string memo (dict stamp → canonical) inside `_chat_messages_to_responses_input` would hoist it trivially.

2. **Canonicalization can *re-classify* a legacy `other:` stamp, not just normalize it.** `_canonicalize_responses_issuer_kind("other:https://githubcopilot.com/...")` would return `github_responses`, since `_classify_responses_issuer` checks the known hosts before the `other:` fallback. That's arguably a correctness win (legacy stamps pointing at a recognized host now match their real category), but it's a behavior the tests don't pin — the canonicalize tests only cover custom endpoints. Worth a one-line test asserting a legacy `other:` stamp that happens to point at a github/xai host re-classifies to the category, so the intent is locked in.

3. **The `msg`-prefix rule encodes an assumption about Codex's id namespace.** `stripped_id.startswith("msg")` means any future valid Codex id that doesn't use the `msg_` prefix gets silently withheld (text preserved, id dropped). That's the safe direction and matches today's reality, and the docstring documents it — just flagging that if Codex ever emits non-`msg` ids, the guard is already positioned to discard them rather than break the turn. A fallback (keep id if `canonical_current_issuer == codex_backend` AND the id was stamped by that same issuer, even without `msg_`) would be more precise, but only matters if Codex changes.

4. **The `_classify_responses_issuer` exception fallback (`str(base_url).rstrip("/")`) leaves an unnormalized stamp.** If `httpx.URL` raises (invalid URL), the fallback stores the raw string, which can differ from the SDK-normalized form for the same endpoint — so a fallback-path stamp and an httpx-path stamp for one endpoint wouldn't match. The tests cover the httpx path thoroughly but not the fallback; a parametrized case with a URL that forces the fallback would close it.

Nice work on `test_responses_issuer_canonicalizes_sdk_base_url` — that's exactly the contract (configured vs SDK-normalized agree) the whole feature depends on, and the `None`/empty-issuer paths are handled cleanly.

Amekn:
Rebased and updated to `f538ae315c126e7dd07876970d7a77834bb5cd88` on current `main` (`67764dc0863349a384c16425e73ee8571f3a94b7`). The branch is now one commit ahead, zero behind, and GitHub reports it mergeable. Original authorship and author date are preserved.

The conflict resolution ports the fix into the current replay helpers and auxiliary request builder, rather than restoring the old inline implementations. Tests are consolidated into two parameterized behavior tests per the current contribution guidance.

Verification:
- Unmodified base: 21 of the 25 new cases fail for the expected behavior, with no collection/runtime errors.
- Rebased fix: **25 new regression cases + 602 related existing tests pass**, with file retries disabled.
- Ruff and git diff checks pass.
- The integration cases exercise actual SDK serialization and HTTP/SSE exchanges against a local server through both main and auxiliary paths. No live provider accounts are used.

[Baseline evidence](https://github.com/Amekn/hermes-agent/actions/runs/34493333442) | [Rebased validation evidence](https://github.com/Amekn/hermes-agent/actions/runs/34493831901)

The full repository suite was not rerun for this revision. The description now links #92311 and explains the additional coverage compared with #92325.

Regarding the earlier automated review: issuer classification is flag-based, so URL canonicalization deliberately does not reclassify a known hostname into a named provider category. The malformed-URL fallback is now covered. The Codex namespace check remains `msg`, not `msg_`.

**Maintainer action remaining:** the new upstream CI, Docker, and Nix runs are all marked `action_required` and need workflow approval. The temporary fork-only validation branches have been removed; no workflow changes are included in this PR.

Amekn:
Hello? Anyone has an update on this PR? I had to reapply the patch for last few weeks due to hermes update overwrite my changes

kshitijk4poor:
Thanks @Amekn. The reported failure (#92311, `Expected an ID that begins with msg`) is fixed on current main: `agent/codex_responses_adapter.py::_assistant_message_item` drops non-`msg` ids when the current issuer is the Codex backend (f7b6a2b59f, cherry-pick of #92325 by @fangliquanflq, the earlier submitter). The remaining part of this PR — stamping issuer provenance on persisted message items and withholding foreign ids for non-Codex issuers — has no reproduced failure behind it, and the branch now conflicts with main's `_canonical_issuer_kind` plumbing. Closing; if you hit a concrete 400 from another Responses provider on a foreign id, a small PR against current main with that repro would be very welcome.

Amekn:
Thanks Ty, just confirmed the fix works

PR #41056, opened 2026-06-07
feat(lsp): add csharp-ls and fsautocomplete via dotnet global tools

pixu-bd:
csharp-ls and fsautocomplete ship as .NET global tools (`dotnet tool install -g <pkg>`), not via npm or pip. Add a `dotnet` install strategy in agent/lsp/install.py that runs the install command and symlinks the resulting binary from `~/.dotnet/tools/<bin>` into the LSP staging dir, so a single `_existing_binary` PATH probe finds it the same way it does for npm/pip installs. The symlink (not a copy) means `dotnet tool update -g` upgrades the binary in place.

Adds server entries in agent/lsp/servers.py and a partner test covering install detection on PATH.

## Related Issue

Fixes #

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- 

## How to Test

1. 
2. 
3. 

### Code

- [ ] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [ ] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [ ] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
- [ ] My PR contains **only** changes related to this fix/feature (no unrelated commits)
- [ ] I've run `pytest tests/ -q` and all tests pass
- [ ] I've added tests for my changes (required for bug fixes, strongly encouraged for features)
- [ ] I've tested on my platform: 

### Documentation & Housekeeping

- [ ] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A
- [ ] I've updated `cli-config.yaml.example` if I added/changed config keys — or N/A
- [ ] I've updated `CONTRIBUTING.md` or `AGENTS.md` if I changed architecture or workflows — or N/A
- [ ] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A
[...]

PR #97247, opened 2026-08-28
feat(lsp): add fsautocomplete F# language server

jmhickman:
Registers Ionide's fsautocomplete for .fs/.fsi/.fsx, discovers the dotnet global tool in ~/.dotnet/tools when it isn't on PATH, and auto-installs via `dotnet tool install --tool-path`.

## What does this PR do?

Currently, Hermes does not include support for F#'s LSP provided by the Ionide project (https://ionide.io/). This PR addresses this shortcoming.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made. 

Added `fsautocomplete` support for reducing the number of turns required for a model to fix or address issues before building the application.

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate - https://github.com/NousResearch/hermes-agent/pull/41056 stalled/abandoned PR
- [x] I've tested on my platform: Fedora Linux 44

- [x] I've updated relevant documentation: `lsp.md`
- [N/A] I've updated `cli-config.yaml.example` if I added/changed config keys 
- [N/A] I've updated `CONTRIBUTING.md` or `AGENTS.md` if I changed architecture or workflows 
- [N/A] I've updated tool descriptions/schemas if I changed tool behavior

### DIFF
--- a/agent/lsp/install.py
+++ b/agent/lsp/install.py
@@ -60,6 +60,10 @@ INSTALL_RECIPES: Dict[str, Dict[str, Any]] = {
     # PowerShellEditorServices is a release-zip bundle driven by pwsh; we probe
     # the host so `hermes lsp status` reports its presence.
     "powershell": _manual("pwsh"),
+    # F# — fsautocomplete is a .NET global tool. Auto-install via
+    # ``dotnet tool install --tool-path``. Also probed in ``~/.dotnet/tools``
+    # because ``dotnet tool install -g`` often leaves that dir off PATH.
+    "fsautocomplete": _recipe("dotnet", "fsautocomplete", "fsautocomplete"),
 }
 
 _install_locks: Dict[str, threading.Lock] = {}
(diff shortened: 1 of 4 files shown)


alt-glitch:

Related to #41056: both add fsautocomplete support, but this PR uses Hermes-owned dotnet tool-path installation and adds F#-specific workspace handling rather than the earlier global-tool approach.

Enough1122:

Solid, well-tested addition. The `dotnet` install strategy, `_dotnet_tools_dir()` probing (with the `DOTNET_CLI_HOME` redirect honored correctly), the `AutomaticWorkspaceInit` requirement, and the state-directory isolation under `<HERMES_HOME>/lsp/` are all coherent, and the test suite is genuinely thorough — the spawn tests, the DOTNET_ROOT not-clobbering test, and the workspace-boundary root tests each pin a real behavior.

A few points to consider:

1. **`_root_fsharp` walks up to 64 levels unbounded when `workspace` is falsy.** `_fsharp_dir_markers` runs `os.listdir` on every directory on the way up, and the loop only stops at the workspace boundary, filesystem root, or the 64-step cap. If a caller ever passes an empty/None workspace (the docstring says "then the git workspace" as if it always exists), a stray `.sln` high in the tree (e.g. a system/global solution under `/` or a drive root) would be picked up as the project root. The existing test pins the workspace-boundary case but only when `workspace` is provided. Consider explicitly returning `workspace`/falling back immediately when `workspace_abs` is None, so the up-walk is always bounded by a real boundary.

2. **`state_dir` uses `sha256(abspath(root))` — Windows case-insensitivity can spawn duplicate cache dirs.** `C:\repo` and `c:\REPO` (same directory, different case) produce different digests, so the same workspace accumulates two state directories. Minor and harmless (just a bit of stale cache), but a `normcase` before hashing would make it collision-stable on Windows, matching the `normcase` discipline the codebase uses elsewhere for path identity.

3. **`Directory.Build.props` / `Directory.Packages.props` as F# root markers are weaker signals than the rest.** They're MSBuild-level files present in many C#-only repositories; since `_root_fsharp` only runs for `.fs`/`.fsi`/`.fsx` files this is usually fine, but a repo where the F# file sits in a subfolder with its own `Directory.Build.props` while a sibling C# solution owns the outer one could mis-root. `paket.dependencies` and `global.json` are the strong F# signals; consider ordering the markers or documenting that the MSBuild files are best-effort. Not blocking — just a note on intent.

4. **Minor: `_install_dotnet` always provisions via `dotnet tool install --tool-path` even when the user already has a global tool.** The docstring says the staging copy is deliberate (Hermes-owned), and `_existing_binary` probes `~/.dotnet/tools` first, so a pre-existing global install is reused — good. Just confirming that's the intended precedence (existing global tool wins over a staged copy), since a reviewer could otherwise expect the reverse.

The F# support is a nice addition and the docs table + install instructions are accurate. Good work.

jmhickman:
So, is there anything else I need to do to push this over the finish line?

jmhickman:
Would still like to get this merged.

jmhickman:
Updated to resolve conflicts.

PR #97249, opened 2026-08-28
fix(prompt-size): report GUI and subagent surfaces accurately

irakli1980h-blip:
## What does this PR do?

`hermes prompt-size` is the tool you reach for when a session's prompt is
bigger than expected. On two surfaces it reported a prompt that no real
session would receive, because it re-derived the toolset list instead of
asking the code that actually assembles it.

- **desktop/tui** — these sessions are not built from a `hermes-<platform>`
  composite. `tui_gateway.server._load_enabled_toolsets` resolves the CLI
  toolsets and then folds in the client-surface toolsets that are
  deliberately kept off `_HERMES_CORE_TOOLS` (`project`, `desktop_ui`).
  The diagnostic never added them.
- **subagents** — `tools/delegate_tool.py` never resolves a
  `hermes-subagent` composite either. A child inherits the parent's
  enabled toolsets and is separately handed
  `_blocked_toolsets_for_role(role) + ["kanban"]` as `disabled_toolsets`,
  so blocked tools are subtracted *after* composite expansion.
- **tool_search bridge** — those tools are synthesised at assembly time by
  `tools.tool_search.assemble_tool_defs` and carry no registry
  attribution, so their bytes were reported as unattributed.

The fix delegates to the runtime resolvers rather than re-deriving the
sets, so the diagnostic stays in lockstep when the runtime changes. A
conservative fallback keeps trimmed installs working when the gateway
module is unavailable.

The accompanying optional skill (`hermes-diagnostic-fidelity`) writes up
the general procedure, since this is a recurring shape: a diagnostic
disagrees with runtime, and the disagreement gets read as a product bug
rather than a reporting bug.

## Related Issue

No existing issue — found while auditing a live install where
`prompt-size` under-reported a desktop session.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] 🎯 New skill (bundled or hub)

## Changes Made

- `hermes_cli/prompt_size.py` — `_resolve_gui_surface_toolsets()` delegates
  to `tui_gateway.server._gui_surface_toolsets`;
  `_resolve_subagent_toolsets()` models a leaf child of a CLI-configured
  parent; deferred bridge tools get their own attribution bucket.
- `tests/hermes_cli/test_prompt_size.py` — 16 tests asserting the
  diagnostic agrees with the runtime resolvers.
- `optional-skills/software-development/hermes-diagnostic-fidelity/SKILL.md`
  plus its docs page and catalog/sidebar registration.
- `tests/skills/test_hermes_diagnostic_fidelity_skill.py` — 12 tests.
- `website/` — docs typecheck was failing on `main` with TS5101
  (`baseUrl` deprecated); fixed alongside the catalog registration,
  isolated in its own commit.

## How to Test

1. `pytest tests/hermes_cli/test_prompt_size.py tests/skills/test_hermes_diagnostic_fidelity_skill.py -q`
   → 28 passed.
2. `hermes prompt-size --platform desktop` — `project` / `desktop_ui` now
   appear in the toolset breakdown; before this change they were absent.
3. `hermes prompt-size --platform subagent` — reports the prompt a real
   delegated child receives (parent's toolsets minus the blocked set),
   instead of an unresolvable composite.
4. `cd website && npx tsc --noEmit -p tsconfig.json` → exit 0 (fails on
   `main` with TS5101).

### Code

- [x] I've run the relevant suites and they pass — see note below
- [x] I've tested on my platform: Debian 12 (aarch64), Python 3.11.15

**Note on the full suite:** `pytest tests/hermes_cli/ -q` reports
`1 failed, 1053 passed` — `test_config.py::TestSanitizeEnvLines::test_migrate_reports_normalized_line_formatting`.
That failure reproduces identically on a clean `origin/main` checkout with
this branch's changes absent, and the test passes in isolation on both —
it is a pre-existing ordering/state-leak issue, not a regression from this
PR. Happy to open a separate issue for it.

- [x] I've updated relevant documentation — skill docs page + catalog +
      sidebar; module docstrings updated in `prompt_size.py`
- [x] N/A — no config keys added or changed
- [x] I've considered cross-platform impact — the GUI resolver is imported
      defensively with a conservative fallback, so trimmed or
      gateway-less installs (any OS) degrade to the documented default
      rather than raising

### DIFF
--- a/hermes_cli/prompt_size.py
+++ b/hermes_cli/prompt_size.py
@@ -32,6 +32,11 @@ _NAMES_ONLY_LINE_RE = re.compile(r"^  .+ \[names only\]: (?P<names>.+)$")
 # Cap the human-readable "Skills by size" table; ``--json`` always has them all.
 _SKILLS_TABLE_LIMIT = 20
 
+# Bucket label for the tool_search bridge tools. They are synthesised at
+# assembly time (``tools.tool_search.assemble_tool_defs``) rather than
+# registered in a toolset, so they have no registry attribution of their own.
+_DEFERRED_BRIDGE_LABEL = "(deferred tool-search bridge)"
+
 
 def _bytes(s: str) -> int:
     return len(s.encode("utf-8"))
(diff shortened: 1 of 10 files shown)


Enough1122:

This is a well-executed fix. The three composite-less platforms (`desktop`/`tui`/`subagent`) genuinely own no `hermes-<platform>` toolset, so a bare `_get_platform_tools` lookup producing an empty set was a real reporting defect — and routing each through its actual host resolver (tui_gateway's `_gui_surface_toolsets` for GUI, delegate_tool's `_strip_blocked_tools`/`_blocked_toolsets_for_role` for subagents) is exactly the "delegate, don't re-derive" discipline the new skill preaches. The two-pass skill-name mapping (declared names first, dir aliases only as fill) fixes a genuine shadowing bug with a precise regression test, and the deferred-bridge bucket un-hides the largest schema lever. Test quality is high throughout (the revert-check discipline, the framing-bytes assertion, the render-truncation test).

A few points:

1. **The `_resolve_gui_surface_toolsets` fallback mirrors the very thing the skill forbids — and it's unlabelled.** When `tui_gateway.server` is unavailable you return a hardcoded `{"project"}` (+ `desktop_ui` for desktop). That's a re-derived copy that can silently drift from the real resolver, and unlike the runtime-import path it carries no marker. Given this PR's own skill makes "label any mirror as a mirror" a completion criterion, consider at minimum logging a warning when the fallback is taken (so the reported numbers are known to be estimates), and keeping the comment that names it a conservative mirror. Right now a trimmed-install user gets `(unknown)`-free but potentially wrong numbers with no signal.

2. **`_resolve_subagent_toolsets` models "leaf child of a CLI parent" and the report will be wrong for other shapes.** The docstring says so, but the user-facing output of `hermes prompt-size --platform subagent` is presented as "a prompt a real subagent would receive." A subagent spawned from a desktop/gateway parent (or an orchestrator-role child) gets different enabled/blocked toolsets, so the printed number is the CLI-parented-leaf estimate only. Worth one line in the command's help or the output itself ("models a leaf child of a CLI-configured parent") so nobody reads it as the universal subagent payload.

3. **Unrelated changes ride along and would be cleaner as their own PR or a stated reason.** The website `tsconfig.json` additions (`ignoreDeprecations`, `resolveJsonModule`, `paths`), the `JSX.Element` → `React.JSX.Element` churn in two components, and the `type: 'ref'` sidebar conversion have nothing to do with "report GUI and subagent surfaces accurately" — they look like build-compat fixes for this PR's docs page. They're small, but a reviewer grepping the diff for the prompt-size behavior has to wade through them. If they're required to keep the site building, a sentence in the PR description would pre-empt the "why is tsconfig changing?" question.

4. Minor: the `BRIDGE_TOOL_NAMES` import is wrapped in try/except with a silent empty-frozenset fallback — so a trimmed install falls back to `(unknown)` for bridge tools again, which is the exact label this PR exists to remove. Same class as #1; a warning on the fallback would make the degradation visible.

The skill itself is well-authored (sections in modern order, checkable completion criteria, the pinned-toolchain pitfall is a genuinely useful lesson), and `test_description_within_index_window` correctly treats the description length as a contract rather than a snapshot. Nice work.

PR #93009, opened 2026-08-23
feat(cron): persist conservative transport receipts

Kinkoolino-Hermes:
## What does this PR do?

Adds a conservative, typed transport-receipt contract for cron delivery so Hermes can distinguish provider-confirmed delivery from ambiguous or definite failure outcomes.

A target is `delivered` only when every preregistered text/media component has an explicit provider acknowledgement. Timeouts, legacy `success + message_id` results, partial acknowledgements, post-ack persistence failures, and interrupted recovery remain `unknown` and never trigger a blind resend, mirror, session seed, or replay seed.

The implementation keeps requested and actual targets separate, binds evidence to the exact job/fire/component/ordinal/content hash, and revalidates the full typed receipt at the SQLite trust boundary before persistence. Public cron surfaces expose bounded summaries; sensitive editable configuration is available only through the authenticated management-detail endpoint for an explicitly selected profile.

## Related Issue

Fixes #70945

Related but not duplicate:

- #75876 addresses only the live-adapter timeout race and explicitly does not implement durable receipts.
- #82782 adds a broader execution-accountability design, but its current delivery evidence is an untyped whole-delivery record, infers provider IDs from legacy `message_id`/`id`, and has no preregistered per-component ordinal/content-hash plan.
- #90293 addresses provider/model fallback atomicity and evidence-poor execution outcomes rather than transport acknowledgements.

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [x] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Add immutable `TransportTarget` / `TransportReceipt` types and additive receipt preservation on `SendResult`.
- Add a normalized schema-v5 SQLite plan/attempt ledger with atomic migration, retention cascade, exact replay binding, partial evidence preservation, and persistence-boundary constructor revalidation.
- Carry durable execution/fire identity through builtin, external-provider, standalone, relay, direct-tool, and background-pool paths; keep immutable nominal `fire_at` separate from mutable claim heartbeat time.
- Preregister complete text/media plans before the first side effect and settle a target only after all planned components are acknowledged.
[...]

PR #75876, opened 2026-08-01
fix(cron): treat live-adapter confirmation timeouts as unknown

Zhou-Ruichen:
## What does this PR do?

Stops cron from using the boolean returned by `concurrent.futures.Future.cancel()` as proof that a live-adapter coroutine never started.

A real `asyncio.run_coroutine_threadsafe()` reproduction shows `cancel() == True` after the delivery coroutine has entered. Cancel-before-loop behavior also differs between Python 3.11.15 and 3.14.6. The boolean is therefore cancellation state, not a dispatch or transport receipt.

After a scheduled live-adapter confirmation timeout, this change performs best-effort cancellation, records `delivery status unknown`, and skips the standalone retry. If scheduling fails before a Future is returned, the existing standalone fallback remains available.

This deliberately favors the no-duplicate invariant for an ambiguous outcome. It does not guarantee delivery or implement durable transport receipts.

## Related Issues and PRs

Related to #38922 and the timeout split introduced by #50018. The real-Future behavior matches the investigation in #51189.

This completes the auditability direction in #70963 for the still-unsafe cancellable branch. It does not close or claim the broader per-target receipt design in #70945.

## Type of Change

- [x] Bug fix
- [x] Tests

## Changes Made

- `cron/scheduler.py`: ignore the cancellation return value, record the timeout as unknown, and skip retry/media/mirroring for that target.
- `tests/cron/test_scheduler.py`: replace the synthetic Future with a real background loop/Future regression and retain coverage for pre-scheduling fallback.

## How to Test

1. Run `scripts/run_tests.sh tests/cron -q`.
2. The timeout regression proves the coroutine entered and the first scheduler-owned `cancel()` returned `True`.
3. Verify the result contains `delivery status unknown` and the standalone sender was not awaited.

## Verification

- `scripts/run_tests.sh tests/cron -q`: 374 passed, 0 failed.
- `scripts/run_tests.sh tests/cron/test_scheduler.py -q`: 64 passed, 0 failed.
- `.venv/bin/ruff check cron/scheduler.py tests/cron/test_scheduler.py`: passed.
- `git diff --check origin/main...HEAD`: passed.
- Full `scripts/run_tests.sh` was attempted but did not complete cleanly in this checkout: the optional ACP extra is not installed (`ModuleNotFoundError: acp`), and unrelated LSP tests hit the live-system signal guard. The focused and package-level cron suites are green.

Tested on macOS, Python 3.11.15.

### Code

- [x] Read the Contributing Guide.
[...]

PR #96894, opened 2026-08-28
fix(cron): mark cron live-adapter deliveries as new conversation rounds

Jackal991:
## Summary

Cron deliveries ride the same `adapter.send()` path as in-turn sends but carry **no round-lifecycle signal**. Round-tracking platform adapters (e.g. the `lightclawbot` plugin from #96876) fall into their "reuse last closed round" branch and append the cron report to the **previous user turn's bubble** — invisible in the user's current conversation while the scheduler logs `delivered ... via live adapter`. Real incident from the issue (2026-08-28): a weekly-review cron report silently vanished this way.

The plugin cannot distinguish "cron delivery" from "post-`stop_typing` attachment follow-up" locally (both are: no active round + `_last_round_id` present) — that ambiguity is exactly why the signal must come from the framework side, which is the fix direction #96876 itself prefers.

## Changes

- `cron/scheduler.py::_deliver_result` stamps a gateway-internal `new_round: True` marker onto **both** the text-routing metadata (`route_metadata`) and the media metadata (`media_metadata`) for cron live-adapter sends. Stamped after the DM-topic routing if/else so both branches carry it.
- Contract: adapters that track rounds use the marker to open a **fresh** round/msgId for these sends instead of reusing the last closed one; adapters that don't track rounds ignore unknown metadata keys, so this is a no-op for every existing platform (verified: `DeliveryRouter._deliver_to_platform` passes metadata through verbatim, no validation/whitelist anywhere in the chain).
- Follows the existing gateway-internal metadata marker precedent (`_interim_send` / `non_conversational` in `gateway/run.py`).
- Standalone fallback path untouched — it builds fresh adapters per send and has no round-lifecycle state, so it is structurally safe.

## Tests

Two regression tests in `tests/cron/test_scheduler.py` (`TestCronDeliveryNewRoundMetadata`), driving the real `_deliver_result` path through the file's existing spy-router harness:

- `test_cron_delivery_metadata_carries_new_round` — the text send's routing metadata carries `new_round: True` (+ `job_id`).
- `test_cron_media_metadata_carries_new_round` — the media send on the same delivery carries the same marker (guards the legitimate post-`stop_typing` attachment case the reuse branch exists for).

**Revert-to-buggy proof**: both tests fail on base `99c3cad857` with the exact assertion signature (`assert None is True` on `metadata.get("new_round")`) and pass with the fix.

```
tests/cron/test_scheduler.py -q            → 105 passed (103 baseline + 2 new), 0 regressions
[...]

Issue #78644, opened 2026-08-04
Shard gateway/platforms/base.py (god-file decomposition)


andrexibiza:
**Type:** Refactoring / Architecture
**Guideline Match:** "Refactor god-files into clean modules." / "Extend, don't duplicate. Design one shared interface."

**Description:**
`gateway/platforms/base.py` is a god file: **6,861 lines** on `main` (`42708f8bb3`). Repo-wide policy (2026-08): **all god files are sharded, never reverted** — the only correct answer is to shard the entire thing. Precedent: `gateway/run.py` (#54962) — sharding SHIPPED via 38 extraction PRs (all individually linked on the epic #78647). Nothing here is a request to revert, shrink-truncate, or abandon the file.

**Proposed Changes:**
- **Graph-gated decomposition (gate 1 first):** model the whole module (not pre-seeded clusters) with double-blind independent trees, merge to one typed graph under agreement provenance (≥4/5 agreed = load-bearing), then quality-gate the graph before any cut.
- **One PR per clean cluster/mixin, in waves:** byte-verbatim moves (docstrings, comments, exclusive constants), identity-preserving re-exports (`<module>.<name>` stays green), `Part of` linkage to this issue, suite-green on current main before open.
- **Zero behavior change.** No reverts. Full shard or untouched — partial "just extract this one class" escapes are not the policy.
- **Test contracts move with the code:** monkeypatch/logger patch targets and source-scanning tests that read this file get their targets updated in the same extraction PR.

**Tracking:** consolidated under the repo-wide god-file epic (meta issue, created with this series).

andrexibiza:
Tracking epic: #78647 — repo-wide god-file sharding (20 files, this file included). Scoreboard updates land here per shipped slice.

andrexibiza:
## 🔒 KILL LOCK — gateway/platforms/base.py (6,861 lines @ pin) — 5/5 DEAD (2026-08-06)

**Anchored to Kill All Gods epic #78647.** The former whole: 6,861-line BasePlatformAdapter — proxy/text/media mixins, document cache, stream rendering, streaming TTS, topic sessions, and the long tail a platform base monolith breeds. The mess it caused lives on the epic's Pantheon ledger (2K-law violation).

### Kill table (5 slices, all SHIPPED — open, individually linked)
| Slice | PR | Status |
|---|---|---|
| s1 proxy/text/media mixins | #79593 | SHIPPED |
[...]

PR #97250, opened 2026-08-28
fix(cron): suppress standalone fallback after adapter reports an ambiguous write

### LINKED ISSUE #97230: cron: suppress fallback after adapter reports an ambiguous write
## Bug

When a live adapter partially delivers a multi-message payload and then returns an explicit ambiguous-write result, cron treats the unconfirmed result like an ordinary failure and replays the entire payload through standalone fallback. Chunks that already landed are duplicated.

## Observed sequence

1. A cron result contains several independently deliverable message chunks.
2. The live adapter confirms the first few chunks.
3. A later chunk may have been accepted, but readback cannot prove the outcome.
4. The adapter returns `SendResult(success=False, retryable=False, ...)` with a durable receipt and a structured ambiguous-write state.
5. `cron/scheduler.py` invokes standalone fallback from chunk one.
6. The user receives duplicate copies of the already-delivered prefix.

This is distinct from a proven pre-send failure: once bytes may be on the wire, replay is unsafe.

## Expected behavior

Hermes should expose and honor a transport-neutral ambiguous-write outcome on `SendResult` (for example, `error_kind="ambiguous"` or an equivalent typed field):

- cron records/logs the unknown outcome and receipt;
- cron does not invoke standalone fallback or any second send path;
- ordinary retryable failures and permanent pre-send refusals keep their existing behavior;
- adapters do not need to lie with `success=True` merely to suppress replay.

Adapter-specific `raw_response` values should not become the core contract.

## Related work

#93629 correctly treats a timeout after gateway-loop dispatch authorization as indeterminate and avoids retrying that timeout path. This issue covers the sibling path where the adapter returns an explicit ambiguous `SendResult` after partial delivery.

A local incident regression proves the failure and the no-fallback behavior for object and dict result shapes: https://github.com/calebhicks/hermes-agent/pull/2. That proof currently keys on an adapter-specific `raw_response` value; the upstream fix should promote ambiguity into the platform-neutral `SendResult` vocabulary instead.

## Scope / non-goals

Smallest complete fix:

- one typed ambiguous-write state;
- scheduler no-fallback handling;
- focused tests for object/dict results and preservation of ordinary fallback.

Do not add a delivery ledger, per-chunk resume worker, new configuration, or adapter-specific branching to cron in this issue. Exact tail resumption can be a separate capability after the at-most-once safety floor exists.

Jackal991:
## What does this PR do?

Teaches the cron scheduler to honor a producer-declared **ambiguous write** on a live-adapter `SendResult` instead of replaying the whole payload through standalone fallback.

When a live adapter partially delivers a multi-chunk payload and then returns an explicit non-retryable ambiguous-write result, `cron/scheduler.py` treated it like any ordinary unconfirmed failure and re-sent the entire payload from chunk one through the standalone path — duplicating the chunks that had already landed (issue #97230; live incident proven in calebhicks/hermes-agent#2).

## Related Issues and PRs

Closes #97230
Related: #93629 (gateway-loop timeout path — sibling trigger, both needed)
Complements (not duplicates): #75876 (confirmation *timeout* race — different trigger: `future.result()` TimeoutError, not an adapter-returned ambiguous `SendResult`); #93009 (durable transport-receipt ledger — a deliberate non-goal of #97230).

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes a bug)
- [x] ✅ Tests (regression coverage added)

## Changes Made

1. **Platform-neutral vocabulary** (`gateway/platforms/base.py`): add `"ambiguous"` to `SEND_ERROR_KINDS` — the existing machine-readable failure categories on `SendResult.error_kind` — with documentation: bytes may or may not be on the wire; replay is unsafe; consumers must record-and-wait, not resend.
2. **Scheduler no-fallback handling** (`cron/scheduler.py`): in the `not send_success` branch of `_deliver_result`, when the failed result carries `error_kind == "ambiguous"` (checked on both the `SendResult` object and dict shapes via the existing shape normalization), the scheduler:
   - logs the unknown outcome at WARNING with the durable receipt preserved;
   - records it in `delivery_errors` (run status stays honest, no "delivered" lie);
   - skips the standalone fallback and any second send path for that target;
   - retryable failures and permanent pre-send refusals keep their existing behavior.
3. **Regression tests** (`tests/cron/test_scheduler.py`): new `TestDeliverResultAmbiguousWrite` covering:
   - object (`SendResult(success=False, retryable=False, error_kind="ambiguous", ...)`) → standalone fallback NOT invoked;
   - dict (`{"success": False, "retryable": False, "error_kind": "ambiguous", ...}`) → same;
   - ordinary unconfirmed failure still falls through to standalone (fallback preservation);
   - `success=True` with `error_kind="ambiguous"` is treated as delivered (error_kind is a failure category).

## How to Test

1. `scripts/run_tests.sh tests/cron/test_scheduler.py -q` — all pass, including the new regression class.
2. Revert the scheduler change (keep the tests): the ambiguous-shape tests FAIL (standalone is awaited — the duplicate), proving the tests catch the bug.

### Code

- [x] Conventional Commits
- [x] Searched open/closed issues and PRs; no open PR fixes this root cause
- [x] Only changes related to this fix (no lockfile churn, no scope creep)
- [x] Added regression coverage
- [x] No new config keys, no telemetry, no adapter-specific branching in cron

### Documentation and Housekeeping

- [x] Docs update N/A — behavior addition to an existing typed contract, documented at the producer/consumer sites
- [x] `cli-config.yaml.example` N/A — no config keys changed

### DIFF
--- a/cron/scheduler.py
+++ b/cron/scheduler.py
@@ -3010,6 +3010,26 @@ def _confirm_adapter_delivery(send_result) -> bool:
     return bool(getattr(send_result, "success"))
 
 
+def _send_result_error_kind(send_result) -> Optional[str]:
+    """Return the platform-neutral ``error_kind`` from a send result, if any.
+
+    Accepts both shapes ``_deliver_to_platform`` can return (a
+    ``SendResult`` object or, when the silence-narration filter drops the
+    message, a plain dict) so callers never substring-match raw provider
+    text.  Unlike :func:`gateway.delivery._send_result_error_kind`, this
+    accepts non-SendResult/non-dict objects (returns ``None``) instead of
+    raising on a missing attribute — a producer that never set the field is
+    simply unclassified.
+    """
+    if send_result is None:
+        return None
+    if isinstance(send_result, dict):
+        kind = send_result.get("error_kind")
+    else:
+        kind = getattr(send_result, "error_kind", None)
+    return str(kind) if kind else None
+
+
 def _is_channel_dm_topic(
     runtime_adapter: Any,
     chat_id: Any,
@@ -3648,6 +3668,36 @@ def _deliver_result(job: dict, content: str, adapters=None, loop=None) -> Option
                                 else:
                                     err = "no response from adapter"
                                     shape = "None"
+                                send_error_kind = _send_result_error_kind(send_result)
+                                if send_error_kind == "ambiguous":
+                                    # #97230: the adapter reports an AMBIGUOUS
+                                    # WRITE — the payload (for a multi-chunk send,
+                                    # possibly only some chunks) may already be on
+                                    # the wire, but the adapter cannot confirm the
+                                    # final outcome and has told us a replay is
+                                    # unsafe.  Re-sending the entire payload via
+                                    # the standalone path would DUPLICATE the
+                                    # chunks that already landed.  Treat this like
+                                    # the #38922 in-flight timeout: record the
+                                    # unknown outcome honestly and suppress the
+                                    # standalone fallback (and any second send
+                                    # path for this target).
+                                    msg = (
+                                        f"live adapter send to {platform_name}:{chat_id} "
+                                        f"returned ambiguous result ({shape}, "
+                                        f"error={err}) — outcome unknown, NOT "
+                                        "re-sending via standalone to avoid "
+                                        "duplicate delivery"
+                                    )
+                                    logger.warning("Job '%s': %s", job["id"], msg)
+                                    # Record the unknown outcome in the RUN
+                                    # status (not just target_errors — that
+                                    # list is folded in only on the paths
+                                    # below, which this `continue` skips): a
+                                    # possibly-unconfirmed delivery must not
+                                    # report ok/delivered.
+                                    delivery_errors.append(msg)
+                                    continue
                                 msg = (
                                     f"live adapter send to {platform_name}:{chat_id} "
                                     f"returned unconfirmed result ({shape}, error={err})"
--- a/gateway/platforms/base.py
+++ b/gateway/platforms/base.py
@@ -2592,6 +2592,13 @@ class SendResult:
 #   not_found     the target chat/thread/message no longer exists.
 #   rate_limited  the platform throttled the send (flood control).
 #   transient     a connection-level failure that is safe to retry.
+#   ambiguous     an ambiguous write: the send may have PARTIALLY or fully
+#                 succeeded on the wire but the adapter cannot confirm the
+#                 final outcome (e.g. a multi-chunk payload delivered some
+#                 chunks before the failure).  Replaying the payload is
+#                 UNSAFE (duplicates already-delivered chunks); consumers
+#                 must record the outcome and wait for human/adjudicated
+#                 reconciliation instead of re-sending.
 #   unknown       classification did not match any known shape.
 SEND_ERROR_KINDS = frozenset(
     {
@@ -2601,6 +2608,7 @@ SEND_ERROR_KINDS = frozenset(
         "not_found",
         "rate_limited",
         "transient",
+        "ambiguous",
         "unknown",
     }
 )
--- a/tests/cron/test_scheduler.py
+++ b/tests/cron/test_scheduler.py
@@ -1,14 +1,17 @@
 """Tests for cron/scheduler.py — origin resolution, delivery routing, and error logging."""
 
+import asyncio
 import contextlib
 import itertools
 import json
 import logging
 import os
+from concurrent.futures import Future
 from unittest.mock import AsyncMock, patch, MagicMock
 
 import pytest
 
+from gateway.config import Platform, PlatformConfig
 from cron.scheduler import (
     SILENT_MARKER,
     _build_job_prompt,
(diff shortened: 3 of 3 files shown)


andrexibiza (commented):
Jackal991 — the core safety decision here is right: once a live adapter explicitly says the write is indeterminate, cron must stop trying alternate transports for that same target. The object/dict coverage and the ordinary-failure control are useful, and this is correctly distinct from the gateway-loop timeout work in #93629 / the older confirmation-race work in #75876.

I reviewed exact head `42e3920dcd4a5ff946bb1a9ee5d03682b09d1e8c`. While I was reviewing, `main` advanced to `a9e72f1b5843467b7cabd4d361b8088832912afe`; the exact merge base is still `306db2776c6b6f1acc85c31c4dabba3263f0e9fd`, so this head is now 1 commit ahead / 1 behind current main. The new main commit is unrelated `read_file` work, but this is no longer a direct-current-main child and should be restacked before final acceptance.

### P1 — the no-replay branch drops the durable receipt that makes an ambiguous write reconcilable

This PR’s body says the scheduler “logs the unknown outcome at WARNING with the durable receipt preserved,” and #97230’s acceptance contract is explicit: cron must **record/log the unknown outcome and receipt**. The implementation at this head does not actually carry a receipt across the boundary.

The new branch in `cron/scheduler.py::_deliver_result()` reduces the result to:

- `_send_result_error_kind(send_result)`; and
- `_send_result_error(send_result)` / the human error text.

It then appends only a string like `returned ambiguous result (..., error=<err>)` to `delivery_errors`. `SendResult` on this head has `success`, `message_id`, `error`, `raw_response`, retry metadata, etc., but no transport-neutral receipt field. The new tests construct both ambiguous object and dict results **without any receipt** and therefore cannot prove receipt survivability.

That matters because the incident proof linked from #97230 carries the reconciliation identity separately from the generic error text. A perfectly valid ambiguous result can be shaped like:

```python
SendResult(
    success=False,
    error="partial batch delivery",
    error_kind="ambiguous",
    raw_response={... durable receipt identity ...},
)
```

At this head, cron will correctly suppress the replay — good — but the durable identity is gone from the warning / `last_delivery_error` path. The operator is left with “outcome unknown” and no object to reconcile. That is exactly the half-state the issue asked us not to create.

Please preserve a **platform-neutral receipt identity** alongside `error_kind="ambiguous"`, and assert it survives into the scheduler’s recorded/logged unknown outcome for both object and dict shapes. I would not have cron inspect adapter-specific `raw_response`; #97230 explicitly rejects making raw provider shapes the core contract.

There is already complementary work in #93009 (Kinkoolino-Hermes) that defines typed `SendResult.receipt` / `receipts` plus the durable transport ledger. #97250 does **not** need to absorb that whole feature — this safety floor is valuable independently — but the two PRs need one receipt vocabulary. Either extract/reuse the smallest shared receipt seam from #93009, or define the minimal neutral receipt token here and have #93009 restack onto it. Please do not leave two incompatible receipt dialects.

Required witness: ambiguous object + ambiguous dict each carry a receipt that is **not embedded in the error string**; standalone fallback is never called; the exact receipt survives in the returned/recorded delivery error (and warning assertion if practical).

### P1 landing/ownership — this adds a second copy of result interpretation instead of one shared owner

Current `gateway/delivery.py` already defines `_send_result_error_kind()` with the same dict-or-`getattr(..., None)` semantics. The new `cron/scheduler.py` helper duplicates it, and its docstring says “unlike `gateway.delivery._send_result_error_kind`, this accepts non-SendResult/non-dict objects … instead of raising”; that distinction is not real — the existing helper also uses `getattr(result, "error_kind", None)` and safely returns `None` for arbitrary objects (including `None`).

This is exactly the wrong place to fork semantics now that `error_kind` is becoming load-bearing mutation-safety state. Please give `SendResult` interpretation one bounded shared owner and import it from cron rather than duplicating it in the scheduler. The same applies to receipt extraction once added: one normalization boundary, not scheduler and delivery slowly learning different shapes.

That also keeps this from regrowing two already-oversized authority surfaces. `gateway/platforms/base.py` has a live decomposition owner in #78644 / repo-wide #78647 (“all god files are sharded, never reverted”), and `cron/scheduler.py` is already >5K lines on current repository references. I’m not asking this bugfix to solve either decomposition campaign; I am asking it not to create a new duplicate result-semantics owner inside the scheduler. The enum/documentation can stay adjacent to `SendResult` until the owning extraction lands, but normalization should be shared now.

### Merge-order / defect-class interlocks

These are complementary, not duplicates, and they need to compose before standalone replay:

- #97230 (calebhicks) is the origin/spec and the local incident proof establishes the receipt-bearing ambiguous-write shape. #97250 is Jackal991’s typed no-fallback implementation of that safety floor.
- #93629 (ciabata-git) is the sibling gateway-loop fallback path: timeout **after dispatch authorization** is indeterminate and must not retry. It changes the same `_deliver_result()` fallback region, so whichever lands second must preserve both no-replay reasons.
- #75876 (Zhou-Ruichen) is the earlier confirmation-timeout race work. It is currently open/non-mergeable and is historical/adjacent, not a substitute for this adapter-returned ambiguity class.
- #93009 (Kinkoolino-Hermes) is the complementary durable transport-receipt/ledger architecture. It should own the richer reconciliation model; #97250 should own only the immediate at-most-once floor, but with compatible receipt semantics.
- #96894 (also Jackal991) is another open `_deliver_result()` change carrying `new_round` metadata through cron live-adapter sends. It is currently non-mergeable. Restacking must preserve that out-of-band round marker while integrating the ambiguity path; neither change should silently erase the other.

Attribution on this head is clean: `lawrence@hydra-flow.co.uk` is mapped to `Jackal991`. Please preserve calebhicks’ #97230/local-proof credit and the sibling contributors above when restacking; this is a convergence problem, not duplicate-work erasure.

### Hosted acceptance is still absent on the exact commit

For exact head `42e3920d…`, [CI 33186242904](https://github.com/NousResearch/hermes-agent/actions/runs/33186242904), [Nix 33186241852](https://github.com/NousResearch/hermes-agent/actions/runs/33186241852), and [Docker 33186241911](https://github.com/NousResearch/hermes-agent/actions/runs/33186241911) all completed `action_required`; the CI run has zero jobs. The local scheduler run in the PR body is useful development evidence, but the sole commit does not yet have a hosted-green acceptance receipt. After the restack/fix, the new exact head needs the applicable hosted checks green.

The important part is already here: **an ambiguous mutation outcome is not permission to replay**. Preserve the receipt that lets us settle that unknown later, put result normalization under one owner, compose the sibling fallback paths, and this becomes a strong, narrow at-most-once safety fix. Nice catch on a genuinely dangerous delivery edge. 🚀

PR #97253, opened 2026-08-28
fix(tui): show status while idle/auto compaction runs (#97239)

### LINKED ISSUE #97239: [Bug]: TUI shows zero feedback while context compaction runs — turn blocks silently for minutes (api_calls=0)
## Summary

With `compression.idle_compact_after_seconds` enabled, a TUI prompt **blocks for minutes with no visible feedback at all** — no reasoning, no status line, no progress — before the actual turn even starts. From the user's chair the agent looks dead. The compaction summary is generated by the **same slow local model** serving the conversation, so a 40-65k-token session takes 4-11 minutes of dead UI.

## Measured (v0.20.6, TUI, local llama.cpp backend at ~55 tok/s)

Session: 44k-64k tokens, `compression.threshold: 0.85`, `idle_compact_after_seconds: 1`.

| Prompt time | What ran before the turn | Block duration | `api_calls` in final turn log |
|---|---|---|---|
| 17:01:10 | Idle compaction of ~64,105 tokens (2 summary attempts) | **651 s** (10.9 min) | 0 |
| 17:36:27 | Idle compaction of ~44,579 tokens | **256 s** (4.3 min) | 0 |

Agent log excerpt (17:36 turn):

```
17:36:27,862 tui prompt accepted
17:36:27,885 Idle compaction: 747s idle >= 1s, ~44,579 tokens > 25,502 floor
17:36:27,889 context compression started: messages=15 tokens=~44,579
17:36:27,922 Auxiliary compression: using custom (qwen3.8-27b)
17:40:02,059 Auxiliary compression: using custom (qwen3.8-27b)   # 2nd attempt
17:40:43,914 context compression done: total_duration_ms=256016
17:40:43,976 Turn ended: reason=interrupted_by_user api_calls=0/100
```

The user interrupted at 256 s because the UI showed nothing. The summary generation itself was a **non-streamed auxiliary call on the primary local model** — the GPU was busy writing a summary the user never sees, while the TUI rendered no "compacting…" state.

## Why this is a bug

1. **Zero TUI feedback.** `compression.progress_notices` (#52995) only routes routine compression statuses to **gateway chat platforms** (`gateway/run.py:_gateway_compression_progress_notices_enabled`). The TUI has no equivalent — there is no status line, no reasoning pane content, nothing, for the entire compaction window. A user cannot distinguish "compacting for 4 minutes" from "agent hung".
2. **Same-model compaction multiplies latency.** With a local model backend, `auxiliary.compression.provider: auto` resolves to the primary model. Compaction of a large session costs minutes at decode speed, and it runs **synchronously in front of the turn** (idle compaction) or at preflight.
3. **No TUI-facing toggle.** The only way to avoid the block is editing `config.yaml` (`idle_compact_after_seconds: 0`); there is no TUI command/setting.

## Expected behavior

- TUI shows a status while compaction runs (e.g. "compacting context… 44k tokens, ~2 min remaining" or at minimum a spinner state), so the user knows the agent is alive.
- Ideally compaction of a local-model session uses a fast auxiliary model when one is configured, or runs at the turn-finish boundary (see #96891) so the user's next prompt is never blocked on it.

## Related

- #96891 — feat(gateway): compact over-threshold sessions when a turn settles (moves compaction off the prompt path)
- #91158 — feat(agent): every compaction names the trigger that fired it
- #93241 — fix(compression): bound large-session compaction
- #79357 — idle_compact_after_seconds never fires in gateway mode (opposite platform, same feature)

## Environment

- Hermes v0.20.6 (v2026.8.27), TUI on Windows 11
- Provider: custom llama.cpp server, Qwen 27B IQ4_XS, ~55 tok/s decode
- `compression: {enabled: true, threshold: 0.85, target_ratio: 0.2, idle_compact_after_seconds: 1, progress_notices: true}`

HexLab98:
## Summary
- Idle/preflight compaction reached the TUI as a generic `lifecycle` status that did not contain `Compacting context`, so the busy bar kept rotating "cogitating"/"running…" with no hint that the turn was blocked on a minutes-long summary.
- Re-tag those in-progress compression lines as `kind="compacting"`, keep that status until `compacted` (no 4s restore), and freeze the FaceTicker verb on "compacting" so the elapsed clock is visible for the whole pause.

Fixes #97239

## Test plan
- [x] `scripts/run_tests.sh tests/tui_gateway/test_compaction_status.py`
- [x] `npx vitest run` in `ui-tui` for `createGatewayEventHandler` + `appChromeBlockedTimers` (compaction cases)
- [ ] TUI: enable `compression.idle_compact_after_seconds`, grow a large session, send a prompt, confirm the status bar shows `compacting…` with a live elapsed duration instead of a silent hang
- [ ] After compaction finishes, confirm the bar returns to the normal busy ticker and the turn proceeds

### DIFF
--- a/agent/conversation_compression.py
+++ b/agent/conversation_compression.py
@@ -96,7 +96,8 @@ _TERMINAL_COMPRESSION_PROVENANCES = frozenset(
 # status as ``kind="compacting"`` (tui_gateway/server.py::_status_update), so
 # drivers like the desktop app can show an explicit "Summarizing…" indicator
 # instead of the transcript appearing to silently reset. Keep the marker phrase
-# intact if you reword COMPACTION_STATUS.
+# intact if you reword COMPACTION_STATUS. Idle/preflight/retry lines do not
+# contain this marker — ``is_compaction_progress_status`` covers those too.
 COMPACTION_STATUS_MARKER = "Compacting context"
 COMPACTION_STATUS = (
     f"🗜️ {COMPACTION_STATUS_MARKER} — summarizing earlier conversation so I can continue..."
@@ -204,6 +205,40 @@ ROUTINE_COMPRESSION_STATUS_SAMPLES = (
         new_ctx=120000, old_ctx=250000
     ),
 )
+
+
+def is_compaction_progress_status(text: str | None) -> bool:
+    """True for in-progress auto-compaction lifecycle lines (not the done edge).
+
+    ``tui_gateway.server._status_update`` re-tags matching ``lifecycle``
+    statuses as ``kind="compacting"`` so TUI and desktop can show a summarizing
+    indicator for the whole pause. Matching only ``COMPACTION_STATUS_MARKER``
+    left idle/preflight/retry lines looking like a hung turn (#97239).
+
+    The terminal ``COMPACTION_DONE_STATUS`` is emitted as ``kind="compacted"``
+    and must not match here.
+    """
+    if not isinstance(text, str):
+        return False
+    body = text.strip()
+    if not body:
+        return False
+    if COMPACTION_STATUS_MARKER in body:
+        return True
+    if body == COMPACTION_DONE_STATUS:
+        return False
+    lowered = body.lower()
+    if "compaction complete" in lowered:
+        return False
+    # Failure-class overflow warning mentions compression but is a blocked
+    # notice, not progress — keep it lifecycle so chat gateways stay loud.
+    if "compression is currently blocked" in lowered:
+        return False
+    return (
+        "compact" in lowered
+        or "compress" in lowered
+        or "context reduced to" in lowered
+    )
 
 
 def _builtin_memory_prompt_snapshot(agent: Any) -> Optional[Tuple[str, str]]:
@@ -5126,6 +5161,7 @@ __all__ = [
     "COMPACTION_STATUS",
     "COMPACTION_DONE_STATUS",
     "COMPACTION_STATUS_MARKER",
+    "is_compaction_progress_status",
     "check_compression_model_feasibility",
     "replay_compression_warning",
     "compress_context",
--- a/tests/tui_gateway/test_compaction_status.py
+++ b/tests/tui_gateway/test_compaction_status.py
@@ -1,9 +1,9 @@
-"""Auto-compaction status re-tagging for the desktop "Summarizing…" indicator.
+"""Auto-compaction status re-tagging for TUI/desktop summarizing indicators.
 
 Auto-compaction reaches the gateway as a generic ``lifecycle`` status. The
-gateway re-tags it as ``kind="compacting"`` so drivers (the desktop app) can
-show an explicit summarizing indicator instead of the transcript appearing to
-silently reset mid-turn.
+gateway re-tags in-progress lines as ``kind="compacting"`` so drivers can
+show an explicit summarizing indicator instead of the turn looking hung
+(#97239). The marker-only match missed idle/preflight/retry wording.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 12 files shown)


Enough1122:

Good fix for #97239. The root cause is real: the marker-only match (`COMPACTION_STATUS_MARKER in body`) missed the idle/preflight/retry wording, so those lines stayed `kind="lifecycle"` and the TUI/desktop showed no summarizing indicator — looking like a hung turn. Moving to `is_compaction_progress_status()` with the explicit negative cases (done edge, "compaction complete", blocked-overflow warning) and a tight test suite (including the full `ROUTINE_COMPRESSION_STATUS_SAMPLES` sweep) is the right shape. The TUI side freezing the verb on "compacting" while keeping the glyph ticking is also clean.

A few points to consider:

1. **The `compacting` flag has no self-clearing timeout — a lost `compacted` event leaves it stuck.** `createGatewayEventHandler.ts` sets `compacting: true` on `kind === 'compacting'` and clears it only on `kind === 'compacted'` (or turn reset / gateway exit). If the completion event is dropped (backend exception mid-compaction, gateway hiccup), the TUI keeps the frozen "compacting" verb indefinitely until the turn ends. Consider a bounded fallback — e.g. clear `compacting` when the next non-compaction `status.update` arrives, or when a `compacting` status is followed by some elapsed window without a `compacted`/`compacting` event. Cheap insurance against a stuck indicator.

2. **`is_compaction_progress_status` is substring-based and will drift with copy.** The broad `"compact" in lowered or "compress" in lowered` fallback is intentional for the known templates, but any future status line that happens to contain "compress"/"compact" (a user-facing message, a quota warning, a settings notice) will be silently re-tagged as `compacting` and freeze the TUI verb. The template sweep test pins today's strings, but it's a change-detector on prose. If you want it to stay substring-based, a safer gate is to match against the known status *template classes* (idle/preflight/retry/routine) rather than a free-text keyword — or at least require the marker *or* an explicit template-class identifier.

3. **`test_is_compaction_progress_status_covers_routine_in_progress_lines` iterates `ROUTINE_COMPRESSION_STATUS_SAMPLES` and asserts every non-done sample returns True.** That's good, but the inverse is weaker — only a handful of negatives are asserted. A reverse contract would be stronger: every non-compaction lifecycle status the gateway can emit (rate-limit, tool errors, connection drops) must return False. A small "known lifecycle samples" list (like the rate-limit string already tested) would guard the false-positive direction without being a snapshot test.

4. Minor: in `tui_gateway/server.py` the `kind === 'compacting'` re-tag happens for *every* lifecycle status whose body passes `is_compaction_progress_status`, including ones that originate from manual `/compact` (already `kind="compressing"`, so unaffected) — fine. Just double-check the "compression is currently blocked" negative case: the test asserts it stays lifecycle, but a blocked notice arguably *should* show a busy indicator if it's mid-turn. Intentional, just confirming it's deliberate.

Nice test coverage overall — the fake-timer test that advances 4001ms and asserts the status stays visible, then drops on `compacted`, is exactly the behavior contract this needs.

kshitijk4poor:
Closing as **implemented on main**.

The content of this PR landed on `main` as commit `3a542bbef4` ("fix(tui): show status while idle/auto compaction runs") — `is_compaction_progress_status()` in `conversation_compression.py`, the widened gateway match in `tui_gateway/server.py::_status_update`, the `compacting` UiState plumbing (`interfaces.ts`, `uiStore.ts`, `createGatewayEventHandler.ts`), and the FaceTicker `verbOverride` freeze are all present on current main; the test files are byte-identical to this branch. Verified against `origin/main@82c95a2a01`.

Thanks @HexLab98 — the fix for #97239 is live.

PR #97254, opened 2026-08-28
fix(video): require image input for i2v-only models

### LINKED ISSUE #97252: video_generate schema does not require image_url for image-only models
## Summary

`video_generate` builds a capability-gated JSON Schema for the configured provider/model. For image-to-video-only models, the generated description says `image_url` is required and the property is exposed, but the schema's `required` array still contains only `prompt`.

This allows the model to emit a schema-valid call without `image_url`; the provider then rejects it at runtime even though the tool already knew the active model could not perform text-to-video.

## Current behavior

For an active model with `modalities: ["image"]`:

```text
properties = [aspect_ratio, duration, image_url, model, prompt, resolution]
required   = [prompt]
description says image_url is REQUIRED
```

The regression is in `_build_dynamic_video_schema()`: it computes `can_i2v=True` and `t2v=False`, but always returns `required: ["prompt"]`.

## Expected behavior

When the resolved active model supports image input but not text-to-video, `image_url` must be included in the JSON Schema `required` array. Dual-modality and text-only models should continue requiring only `prompt`.

## Regression

Introduced by the capability-gated dynamic schema change in #97095. Existing coverage checks that `image_url` is present and the description says image-to-video-only, but does not assert the schema requirement.

Liuzikaii:
## Summary

- require `image_url` in the generated `video_generate` schema when the resolved model supports image input but not text-to-video
- keep `image_url` optional for dual-modality models
- add regression coverage for both schema contracts

Fixes #97252

## Testing

- `python -m pytest -p no:cacheprovider tests/tools/test_video_generate_schema.py tests/tools/test_video_generation_dynamic_schema.py tests/tools/test_video_generation_dispatch.py -q`
- `python -m ruff check tools/video_generation_tool.py tests/tools/test_video_generate_schema.py`

### DIFF
--- a/tests/tools/test_video_generate_schema.py
+++ b/tests/tools/test_video_generate_schema.py
@@ -229,7 +229,19 @@ class TestDynamicParamGating(unittest.TestCase):
             model_meta={"modalities": ["image"]},
         )
         self.assertIn("image_url", schema["parameters"]["properties"])
+        self.assertEqual(
+            schema["parameters"]["required"], ["prompt", "image_url"]
+        )
         self.assertIn("image-to-video only", schema["description"])
+
+    def test_dual_modality_model_keeps_image_url_optional(self):
+        schema = self._schema_with({
+            "modalities": ["text", "image"], "max_reference_images": 0,
+            "supports_audio": False, "supports_negative_prompt": False,
+            "supports_seed": False, "supports_upscale": False,
+        })
+        self.assertIn("image_url", schema["parameters"]["properties"])
+        self.assertEqual(schema["parameters"]["required"], ["prompt"])
 
     def test_no_provider_serves_prompt_only(self):
         with patch.object(vt, "_resolve_active_provider", return_value=None):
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Makes `image_url` required for image-to-video-only models, optional for dual-modality.

**What it does**
- `tools/video_generation_tool.py:_build_dynamic_video_schema` now builds `required = ["prompt"] + (["image_url"] if can_i2v and not t2v else [])`; for dual `text+image` keeps `required:["prompt"]` with `image_url` optional.
- Tests `test_i2v_only_model_overrides_backend_union` and `test_dual_modality_model_keeps_image_url_optional` pin both cases.

**Non-blocking notes**
- `can_i2v and not t2v` correctly identifies i2v-only vs dual; if model advertises `modalities: ["image"]` with `max_reference_images>0` still counts as i2v — correct.
- Schema description includes "image-to-video only" — already covered.

Minimal schema correctness fix.

Issue #58936, opened 2026-07-05
WhatsApp bridge (Baileys) crashes silently, loses creds.json, requires full re-pairing


webtecnica:
## Description

The WhatsApp bridge (Baileys) crashes silently and loses the session (`creds.json`), requiring full re-pairing. The bridge.log is consistently **0 bytes**, making post-mortem diagnosis impossible.

## Timeline of a Disconnection Event

From a real incident on 2026-07-04/05:

| Time | Event |
|---|---|
| 23:44 (D-1) | First `[Whatsapp] Send failed: Server disconnected — trying plain-text fallback` |
| 00:18 (D-1) | Fallback send also fails — bridge is dead |
| 01:26 | Cron deliveries fail: `"Not connected to WhatsApp"` |
| 03:47 | Bridge restarts, says `Bridge ready (status: connected)` but immediately dies with massive `[Whatsapp] Poll error:` spam (bridge can't connect to itself on port 3000) |
| 09:52-10:08 | Gateway crash-restart loop, bridge repeatedly times out on connect |
| 10:09 | **`creds.json` is gone** — gateway reports `"not paired (no creds.json)"` |

## Root Cause Analysis

The Node.js Baileys bridge process crashes silently:

1. The bridge starts and reports `Bridge ready (status: connected)`
2. It then dies immediately (before the Python adapter can connect to port 3000)
3. No error is logged (bridge.log = 0 bytes)
4. When the gateway recycles, `creds.json` is lost — either deleted by Baileys cleanup during crash, or corrupted during the restart loop

This pattern repeated across **7+ gateway restarts** in a single day, always with the same `Bridge ready → dead → poll error` cycle.

## Impact

- Complete WhatsApp service loss until manual re-pairing
- All scheduled cron deliveries fail silently
- No automatic recovery possible — requires human intervention with `hermes whatsapp`

### 1. Capture bridge stderr (critical)
The bridge.js process should redirect its stderr to a persistent log file, not just `/dev/null` or inherited stderr. The current `bridge.log` is always 0 bytes, making debugging impossible. A simple `fs.writeFileSync` wrapper around stderr would help.

### 2. Backup creds.json before gateway restart
The gateway's shutdown sequence should snapshot `creds.json` to a backup location before killing the bridge process. On startup, if `creds.json` is missing but a backup exists, restore it and attempt reconnection instead of requiring full re-pairing.

### 3. Handle DisconnectReason gracefully
[...]

PR #97255, opened 2026-08-28
fix(whatsapp): keep the bridge alive through listener-less media stream errors

### LINKED ISSUE #97108: [Bug]: WhatsApp bridge process can still die on media-download stream 'error' events that escape saveMedia's try/catch (remaining half of #59261)
### Bug Description

PR #59261 ("contain and surface inbound media download failures") contained the *awaited-rejection* failure mode of inbound media downloads: `saveMedia()` in `scripts/whatsapp-bridge/bridge_helpers.js` now wraps `await downloadMedia(msg)` in try/catch, records the failure, and appends a `[<type> could not be downloaded]` note instead of letting one bad attachment kill the whole `messages.upsert` batch.

There is a second failure mode that this guard cannot reach. `downloadMediaMessage()` (Baileys `7.0.0-rc13`, pinned at `scripts/whatsapp-bridge/package.json:11`) consumes an HTTP stream internally. Under undici/HTTP2 transport, the underlying `Readable` can emit `'error'` **outside the awaited promise chain** — e.g. the socket errors after the buffering consumer has detached, or the decrypt transform's source errors on a later tick. An `'error'` event on an emitter with no `'error'` listener is raised at the event-loop level (`Unhandled 'error' event` / `ERR_UNHANDLED_ERROR`), which:

- is unreachable by the try/catch around the `await` (`bridge_helpers.js:330–352`), and
- becomes an uncaught exception, and the bridge registers **no** `process.on('uncaughtException')` or `process.on('unhandledRejection')` handler anywhere in `scripts/whatsapp-bridge/` (verified at `5fc308a7`), and attaches no `.on('error')` listener to any media stream (zero matches in `bridge.js` / `bridge_helpers.js`),

so the Node bridge process exits. The inbound queue is in-memory (`bridge.js:271` `const messageQueue = []`, filled at `bridge.js:763`), so the in-flight inbound message — and anything else queued but not yet polled by the Python adapter — is lost with no replay when the bridge restarts.

Observed on a production WhatsApp gateway (v0.20.x, August 2026): Baileys media download over undici HTTP/2 emitted `'error'` on a listener-less `Readable`; the bridge process exited mid-turn and the inbound message was permanently lost. This is also consistent with the "bridge dies silently with nothing in bridge.log" family reported in #58936.

### Steps to Reproduce

The real trigger is network-timing dependent (CDN connection torn down mid-transfer), so a deterministic simulation is the practical repro:

1. Run the WhatsApp gateway with media inbound enabled.
2. In `bridge.js`, stub the `downloadMedia` closure (`bridge.js:729`) to model the undici behaviour: resolve the buffer, then on a later tick emit `'error'` on a listener-less stream, e.g. `process.nextTick(() => new (require('stream').Readable)({ read() {} }).emit('error', new Error('HTTP/2 stream reset')))`.
3. Send the bot an image or voice note.
4. The bridge process exits with `Unhandled 'error' event`; the message (and any queued siblings) is gone. Organically, the same exit is triggered by sending large media over a flaky uplink until a transfer is reset mid-stream.

### Expected Behavior

A failed media download — however the failure surfaces — is contained the way #59261 contains rejections: the event carries a `[<type> could not be downloaded]` note, the bridge process stays alive, and no queued inbound message is lost.

### Actual Behavior

The bridge process exits on the unhandled `'error'` event. The in-flight inbound message and the in-memory queue are lost; WhatsApp does not redeliver; the operator sees only a dead bridge (often with an empty bridge.log, cf. #58936).

### Affected Component

Gateway (Telegram/Discord/Slack/WhatsApp) — WhatsApp Baileys bridge (Node side).

### Version

Verified at tag `v2026.8.27` (0.20.6, commit `5fc308a7`):
- `scripts/whatsapp-bridge/bridge.js:729` — `downloadMediaMessage(mediaMsg, 'buffer', {}, { logger, reuploadRequest: sock.updateMediaMessage })`
- `scripts/whatsapp-bridge/bridge_helpers.js:330–352` — `saveMedia` try/catch (the #59261 guard)
- `scripts/whatsapp-bridge/bridge.js:271`, `:763` — in-memory `messageQueue`
- No `uncaughtException` / `unhandledRejection` / stream `.on('error')` handlers in `scripts/whatsapp-bridge/*.js`

Observed in production on v0.20.x. Baileys pin: `@whiskeysockets/baileys 7.0.0-rc13`.

### Notes on fix direction

1. **Last-resort process guards in `bridge.js`** — `process.on('uncaughtException')` + `process.on('unhandledRejection')` that log the error to the bridge log and keep the process alive for non-fatal emitter errors (or exit only after flushing queue state). A guard of this shape has run on our production gateway since August 2026 with no side effects, and the logging half would also address #58936's "bridge.log is always 0 bytes" post-mortem gap.
2. **Attach an error listener where the stream is consumed** — wrap the `downloadMediaMessage` call at `bridge.js:729` so consumption always has an `'error'` listener (e.g. request `'stream'` type and buffer it locally with explicit `stream.on('error', …)`), instead of relying on Baileys' internal `'buffer'` consumption.
3. Worth an upstream note to Baileys as well, but the bridge should not be one unhandled emitter event away from losing mail regardless of what Baileys does.

Related: #59261 (merged; contains the awaited-rejection half), #58936 (open; silent bridge-death family).

liuhao1024:
## What does this PR do?

Contains the second failure mode of inbound WhatsApp media downloads: a Baileys media stream under undici/HTTP2 can emit `'error'` on a listener-less `Readable` **outside the awaited promise chain**, so the `saveMedia()` try/catch from #59261 never sees it. With no `uncaughtException` handler the throw killed the bridge process, and the in-memory inbound queue (`bridge.js` `messageQueue`) was lost with no redelivery.

The fix installs last-resort process guards (`uncaughtException` + `unhandledRejection`) that log through pino and keep the process alive. Keeping the process alive is deliberate and scoped: the failed emitter is already gone, the rest of the bridge is unaffected (the issue reports this exact guard shape running in production since August 2026 with no side effects), and the alternative — losing every queued inbound message — is unrecoverable because WhatsApp does not redeliver. Routing the guard output through pino also lands these failures in the adapter-captured `bridge.log`, closing the "bridge dies silently with an empty log" post-mortem gap from #58936.

The guards live in `bridge_helpers.js` (pure, injectable `log`/`target`) and are wired in `bridge.js` right after the logger exists — the same placement pattern the reconnect/version guards already use.

**Why a consumption-site listener cannot fix this** (evidence from the issue author, see [their comment](https://github.com/NousResearch/hermes-agent/pull/97255#issuecomment-5529095377)): in the pinned Baileys 7.0.0-rc13, `getHttpStream()` returns the undici body as a `Readable` (`lib/Utils/messages-media.js:297`) and `downloadEncryptedContent()` does `return fetched.pipe(output)` (`:515`) without ever exposing `fetched` — and `pipe()` does not forward `'error'` from the source. The source stream is listener-less **by construction**: there is no consumption-site handle the bridge could attach a listener to. A process-level guard is therefore the only fix available from the bridge side; the alternative would have to land upstream in Baileys.

## Related Issue

Fixes #97108

## Type of Change

## Changes Made

- `scripts/whatsapp-bridge/bridge_helpers.js` — new `installProcessGuards({ log, target })`: registers `uncaughtException`/`unhandledRejection` handlers that log the full stack with a kept-alive marker; the log call itself is guarded so a broken log sink can never turn the guard into the killer.
- `scripts/whatsapp-bridge/bridge.js` — import + one wiring line: `installProcessGuards({ log: (line) => logger.warn(line) })` installed right after the pino logger is created, before anything can throw at the event-loop level.
- `scripts/whatsapp-bridge/bridge.process_guard.test.mjs` — new tests (details below).

## How to Test

1. `cd scripts/whatsapp-bridge && node --test bridge.process_guard.test.mjs` — Observed result: `6 passed`. This includes:
   - unit tests: both handler kinds log the full stack with the kept-alive marker; non-Error reasons are stringified; a throwing log sink cannot crash the guard;
   - **end-to-end**: a real child process reproduces the issue's deterministic simulation (`nextTick` a listener-less `Readable.emit('error', ...)` after the awaited download resolves) and prints `ALIVE_AFTER_UNHANDLED_STREAM_ERROR` — the process survives;
   - **negative control**: the identical child script *without* the guards exits non-zero (Node's unhandled `'error'` event behavior) — proving the test pins the real bug mechanism, not a tautology.
2. `node --test bridge.reconnect.test.mjs allowlist.test.mjs outbound_ids.test.mjs owner_message_gate.test.mjs bridge.sendqueue.test.mjs` — Observed result: `21 passed` (no regressions in the sibling suites).
   (`bridge.native.test.mjs` fails identically on a clean checkout of `main` in this sandbox — `ERR_MODULE_NOT_FOUND` because `node_modules` isn't installed here — so it is environmental, not caused by this diff.)
3. Ran `node --check` on both modified production files — both pass.

### Code

- [x] I've run `pytest tests/ -q` and all tests pass — N/A for this change (Node-side bridge only; the repo's Python suite doesn't cover `scripts/whatsapp-bridge/`)
- [x] I've tested on my platform: macOS 26.5 (Node v22)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A (behavior-preserving guard; the function docstring documents the failure mode and rationale)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — N/A (Node `process` events are cross-platform)

## For New Skills

N/A — not a skill.

## Screenshots / Logs

```
$ node --test bridge.process_guard.test.mjs
# pass 6
# fail 0

# guarded child (end-to-end) stdout:
ALIVE_AFTER_UNHANDLED_STREAM_ERROR

# negative control (no guards): child exits non-zero on
# "Unhandled 'error' event" — the pre-fix behavior.
```

### DIFF
--- a/scripts/whatsapp-bridge/bridge.js
+++ b/scripts/whatsapp-bridge/bridge.js
@@ -37,6 +37,7 @@ import {
   buildPollPayload,
   createReconnectScheduler,
   createVersionResolver,
+  installProcessGuards,
   buildLocationPayload,
   buildTextSendPayload,
   createBoundedMessageStore,
@@ -266,6 +267,12 @@ function buildLidMap() {
 let lidToPhone = buildLidMap();
 
 const logger = pino({ level: 'warn' });
+
+// Install before anything can throw at the event-loop level: a listener-less
+// 'error' from a Baileys media stream must not kill the bridge and lose the
+// in-memory inbound queue (#97108). Route guard output through pino so it
+// lands in the adapter-captured bridge log (#58936).
+installProcessGuards({ log: (line) => logger.warn(line) });
 
 // Message queue for polling
 const messageQueue = [];
--- a/scripts/whatsapp-bridge/bridge.process_guard.test.mjs
+++ b/scripts/whatsapp-bridge/bridge.process_guard.test.mjs
@@ -1,0 +1,138 @@
+/**
+ * Unit + end-to-end tests for the last-resort process guards (#97108).
+ *
+ * Regression family: saveMedia()'s try/catch (the #59261 guard) contains the
+ * awaited-rejection half of inbound media download failures, but a Baileys
+ * media stream under undici/HTTP2 can emit 'error' with no listener outside
+ * the awaited promise chain. Without an uncaughtException handler that throw
+ * kills the bridge process and loses the in-memory inbound queue — WhatsApp
+ * does not redeliver. These tests pin both halves: the guard's logging
+ * behavior, and — via a real child process reproducing the issue's
+ * deterministic simulation — that a listener-less stream 'error' no longer
+ * kills the process once the guards are installed.
+ *
+ * These tests avoid importing bridge.js (it starts an HTTP server and a
+ * Baileys socket at module load, same reason as bridge.reconnect.test.mjs);
+ * the guards live in bridge_helpers.js for exactly that reason.
+ */
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { EventEmitter } from 'node:events';
+import { execFileSync } from 'node:child_process';
+import { fileURLToPath } from 'node:url';
+import path from 'node:path';
+
+import { installProcessGuards } from './bridge_helpers.js';
+
+const here = path.dirname(fileURLToPath(import.meta.url));
+
+test('installProcessGuards logs uncaught exceptions and keeps the target alive', () => {
+  const target = new EventEmitter();
+  const lines = [];
+  installProcessGuards({ log: (l) => lines.push(l), target });
+
+  const boom = new Error('HTTP/2 stream reset');
+  target.emit('uncaughtException', boom);
+
+  assert.equal(lines.length, 1);
+  assert.match(lines[0], /uncaught exception contained — bridge kept alive/);
+  assert.match(lines[0], /HTTP\/2 stream reset/);
+  assert.ok(lines[0].includes(boom.stack), 'full stack lands in the log');
+});
+
+test('installProcessGuards logs unhandled rejections too', () => {
+  const target = new EventEmitter();
+  const lines = [];
+  installProcessGuards({ log: (l) => lines.push(l), target });
+
+  target.emit('unhandledRejection', new Error('socket hang up'));
+
+  assert.equal(lines.length, 1);
+  assert.match(lines[0], /unhandled rejection contained — bridge kept alive/);
+});
+
+test('non-Error rejection reasons are stringified, not dereferenced', () => {
+  const target = new EventEmitter();
+  const lines = [];
+  installProcessGuards({ log: (l) => lines.push(l), target });
+
+  target.emit('unhandledRejection', 'plain string reason');
+
+  assert.equal(lines.length, 1);
+  assert.match(lines[0], /plain string reason/);
+});
+
+test('a broken log sink never turns the guard into the killer', () => {
+  const target = new EventEmitter();
+  installProcessGuards({
+    log: () => {
+      throw new Error('log pipeline is broken');
+    },
+    target,
+  });
+
+  // The guard must swallow its own logging failure instead of rethrowing.
+  target.emit('uncaughtException', new Error('original failure'));
+});
+
+// Real-process end-to-end anchors. The child script reproduces the issue's
+// deterministic simulation: resolve the awaited download, then on a later
+// tick emit 'error' on a listener-less Readable — the exact escape hatch the
+// saveMedia() try/catch cannot reach.
+
+const guardedChild = `
+import { Readable } from 'node:stream';
+import { installProcessGuards } from './bridge_helpers.js';
+const lines = [];
+installProcessGuards({ log: (l) => lines.push(l), target: process });
+process.on('exit', (code) => {
+  if (code !== 0) return;
+  if (lines.length < 1 || !lines[0].includes('kept alive')) process.exitCode = 2;
+});
+process.nextTick(() => {
+  // Listener-less 'error' outside any awaited promise chain (#97108 repro).
+  new Readable({ read() {} }).emit('error', new Error('HTTP/2 stream reset'));
+});
+setTimeout(() => {
+  // Reaching this timer at all proves the process survived the throw.
+  console.log('ALIVE_AFTER_UNHANDLED_STREAM_ERROR');
+}, 200);
+`;
+
+const bareChild = `
+import { Readable } from 'node:stream';
+process.nextTick(() => {
+  new Readable({ read() {} }).emit('error', new Error('HTTP/2 stream reset'));
+});
+setTimeout(() => {
+  console.log('SHOULD_NOT_GET_HERE');
+}, 200);
+`;
+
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** Keeps WhatsApp bridge alive through listener-less Baileys media stream errors (#97108).

**What it does**
- `scripts/whatsapp-bridge/bridge.js` installs `installProcessGuards({log:(l)=>logger.warn(l)})` before event loop start, routing `uncaughtException`/`unhandledRejection` from Baileys under undici/HTTP2 through `pino` (adapter-captured bridge log, #58936) instead of crashing and losing in-memory inbound queue (WhatsApp no redeliver).
- `bridge_helpers.js:installProcessGuards` logs "uncaught exception contained — bridge kept alive" with stack, stringifies non-Error rejections, and guards log sink failures.

**Non-blocking notes**
- Guard is last-resort; correct that `saveMedia()`'s try/catch covers awaited half but not listener-less `'error'` emission.
- `target` defaults to `process` but tests inject `EventEmitter` — good seam; verify production `target=process` not double-installed on bridge restart.

Reliability improvement with E2E child process reproduction.

liuhao1024:
Thanks for the review. On the double-install note: the install call runs once at module init (before the event loop starts), and a WhatsApp 515 "restart" is a reconnect, not a process restart — there's no hot-reload path that re-executes the module, so `target=process` never gets a second pair of listeners. The EventEmitter-injected seam in tests is what keeps that property checkable.

adnankassem:
Issue author here — this is the right shape, and I'd like to add one piece of evidence that the PR body doesn't have, because it closes off the "just attach a listener where we consume the stream" alternative that a reviewer might reasonably suggest.

In the pinned Baileys (7.0.0-rc13), `getHttpStream()` returns the undici body as a `Readable` (`lib/Utils/messages-media.js:297`) and `downloadEncryptedContent()` does `return fetched.pipe(output)` (`:515`) without ever exposing `fetched`. `pipe()` doesn't forward `'error'` from the source, so the source stream's error is listener-less **by construction** — there is no consumption-site handle the bridge could attach to. A process-level guard, as done here, is the only fix available from the bridge side; the alternative is upstream in Baileys.

Re-checked upstream `main` at `63279301` (2026-09-03): `scripts/whatsapp-bridge/*.js` still has no `uncaughtException` / `unhandledRejection` handler, so the PR is still needed as-is. We've been running an equivalent guard on a production WhatsApp gateway; no regressions observed.

liuhao1024:
Thanks @adnankassem — this closes off exactly the alternative a reviewer would reasonably reach for. I've folded your Baileys evidence into the PR description (the "Why a consumption-site listener cannot fix this" section): in the pinned 7.0.0-rc13, `downloadEncryptedContent()` returns `fetched.pipe(output)` without ever exposing `fetched`, and `pipe()` does not forward the source's `'error'` — so the stream is listener-less **by construction** and there is no consumption-site handle to attach a listener to. A process-level guard is the only bridge-side fix available; anything narrower would have to land upstream in Baileys. Also good to know the equivalent guard has been running in your production gateway with no regressions — that matches the report in #97108.

PR #97258, opened 2026-08-28
fix(mcp): treat silent ping drop as unsupported rather than dead transport (Closes #97245)

### LINKED ISSUE #97245: MCP keepalive respawn-loops a healthy stdio server that never answers ping (silent drop ≠ -32601)
**Hermes version:** v0.20.5 (2026.8.19), upstream `36b0a96d`; code path unchanged on `origin/main` as of `2026-08-28`
**Platform:** macOS 26.6.2 (arm64), Python 3.11
**Component:** `tools/mcp_tool.py` — `_keepalive_probe()`

## Summary

`_keepalive_probe()` treats a server that **never answers** `ping` the same as a
dead transport, so a healthy stdio server that simply doesn't implement `ping`
gets torn down and respawned on every keepalive tick, forever.

The existing handling is careful about the *legal* shape of "no ping support" —
a JSON-RPC `-32601` latches `_ping_unsupported` and falls back to `list_tools`
(`tools/mcp_tool.py:2862-2884`, hardened in #50028 for servers that phrase it as
`Unknown method: ...`). But a server that returns **nothing at all** never
produces an exception `_is_method_not_found_error()` can classify. It produces a
bare `TimeoutError` after the hardcoded 30s, which is (correctly, for a real
hang) treated as a liveness failure → `mark_suspect` → reconnect.

For a stdio server the reconnect closes the `stdio_client` context, which closes
the child's stdin, so the perfectly healthy subprocess exits on EOF and is
respawned. Next tick, same thing.

Measured on one machine in a single day, from **one** such server:

```
480  keepalive failures
362  respawns  ("starting MCP server 'leann-server'" banners)
```

That was the noisiest process on the box — ~14 failures/hour, flat, around the
clock, independent of load.

## Repro

Any stdio MCP server that answers `initialize` and `tools/list` but never
responds to `ping`. A real-world one is `leann-core` 0.3.7 (`leann_mcp`), whose
dispatcher is an `if/elif` chain with no `ping` branch and no terminal `else`, so
`handle_request()` returns `None` and its `main()` writes nothing — filed
upstream as StarTrail-org/LEANN#403. A ten-line fake server reproduces it just as
well:

```python
import json, sys
for line in sys.stdin:
    req = json.loads(line)
    m = req.get("method")
    if m == "initialize":
        print(json.dumps({"jsonrpc":"2.0","id":req["id"],"result":{
            "protocolVersion":"2024-11-05","capabilities":{"tools":{}},
            "serverInfo":{"name":"mute","version":"1"}}}), flush=True)
    elif m == "tools/list":
        print(json.dumps({"jsonrpc":"2.0","id":req["id"],
                          "result":{"tools":[]}}), flush=True)
    # every other method, ping included, falls through: no response written
```

Register it as a stdio server and watch `~/.hermes/logs/mcp-stderr.log`: a
`starting MCP server` banner every keepalive interval, forever, while the
process is provably healthy (`initialize` and `tools/list` both answer in
~0.0s when driven by hand).

Direct probe, to show the server is not wedged:

```
initialize: {"jsonrpc": "2.0", "id": 1, "result": {...}}          (0.00s)
tools/list: {"jsonrpc": "2.0", "id": 2, "result": {"tools": [...]}} (0.00s)
ping:       *** NO RESPONSE after 15s ***
process still alive: True
```

## Why the existing knobs don't cover it

- **`keepalive_interval`** only changes *how often* the loop runs. It is a
  frequency dial, not an opt-out: any finite value still tears the session down
  on the next tick. Setting it to `3600` is what I ended up doing, and it merely
  converts 362 respawns/day into 24 — the connection still dies hourly for no
  reason.
- **`_ping_unsupported`** is the right latch but is unreachable here, because it
  is gated on `_is_method_not_found_error(exc)` and a silent server yields
  `TimeoutError`.
- **The probe method is not configurable.** The only `_config.get(...)` keys the
  server class reads are `enabled` and `keepalive_interval` (verified on
  `origin/main`), so there is no way to say "use `list_tools` for this server."

## Related but distinct

- **#94335** (closed, fixed) — inverted `_stdio_children_dead()`. Different
  failure: that one fast-fails calls against a *live* child. This one kills a
  live child.
- **#65787** (closed) — moved the probe *from* `list_tools` *to* `ping` because
  `list_tools` is O(tool-count). That change is what exposes this: `ping` is
  cheaper, but it is also **optional**, so the probe now depends on a method
  servers are free not to implement — and some implement it by saying nothing.
- **#64073** — HTTP/Streamable server, ping *times out* on a ~600s cadence. Same
  symptom class, different transport and cause (that server does answer ping,
  intermittently).

## Suggested fix

Either would resolve it; the first is the smaller change.

**1. Treat a first-probe timeout as "ping may be unsupported" and confirm with
the fallback, instead of assuming death.** On the *first* `TimeoutError` from
`send_ping()`, try `list_tools` once before declaring the session dead. If
`list_tools` answers promptly, the transport is demonstrably alive — latch
`_ping_unsupported` and use the fallback from then on. Only if *both* probes
fail is it a genuine liveness failure. This keeps the cheap path for the 99%
and costs one extra round trip, once per connection, for the broken minority.

**2. Make the probe method configurable**, e.g.
`keepalive_probe: ping | list_tools | none`, so a known-mute server can be
pinned to the fallback (or opted out) without waiting on the vendor. This also
gives a workaround for #64073-style servers.

A defensive tweak worth considering alongside either: the probe timeout is
hardcoded `30.0` at both call sites (`:2864`, `:2884`). A server that is merely
slow on a loaded box is indistinguishable from one that is mute. Deriving it
from `connect_timeout`, or capping it at `keepalive_interval`, would avoid
pathological overlap when the interval is short.

I'm happy to put up a PR for option 1 if that's the direction you'd prefer.

Synxneuos:
## Summary
Closes #97245

When an MCP stdio server does not implement the optional `ping` utility and silently drops ping requests (returning no response rather than a JSON-RPC `-32601 Method not found` error, e.g., `leann-core` / `leann_mcp`), `_keepalive_probe()` received a `TimeoutError` and assumed a dead transport, triggering a reconnect cycle on every keepalive tick and causing continuous respawn loops against an otherwise healthy subprocess.

## Changes
- In `tools/mcp_tool.py` (`_keepalive_probe`):
  - On `TimeoutError` / `asyncio.TimeoutError` from `send_ping()`, if the server advertises tool capability (`self._advertises_tools()`), attempt `list_tools()` as a confirmation probe before declaring the connection dead.
  - If `list_tools()` succeeds, the transport is provably alive: latch `_ping_unsupported = True` and log that the server silently drops ping, avoiding both the 30s timeout and the reconnect loop on subsequent ticks.
  - If `list_tools()` also fails, propagate the failure as a genuine liveness breakdown.
- In `tests/tools/test_mcp_capability_gating.py`:
  - Added unit test `test_silent_ping_drop_falls_back_to_list_tools` verifying fallback latching on silent ping drop.
  - Added unit test `test_silent_ping_drop_both_fail_propagates` verifying genuine liveness failure propagation when both probes fail.
  - Added unit test `test_silent_ping_drop_no_tools_propagates` verifying immediate propagation when no tools capability is advertised.

## Verification
- Ran `pytest tests/tools/test_mcp_capability_gating.py -v`: **19/19 passed cleanly in 9.68s**.

### DIFF
--- a/tests/tools/test_mcp_capability_gating.py
+++ b/tests/tools/test_mcp_capability_gating.py
@@ -295,4 +295,54 @@ class TestKeepaliveProbeFallback:
 
         assert task._ping_unsupported is False
 
+    async def test_silent_ping_drop_falls_back_to_list_tools(self):
+        """Regression for #97245: a server that silently drops ping (no
+        response at all) produces a TimeoutError. If list_tools succeeds,
+        the transport is alive — latch _ping_unsupported and return
+        normally instead of reconnect-looping."""
+        task = MCPServerTask("test")
+        task.initialize_result = _caps(tools=SimpleNamespace())
+        task.session = SimpleNamespace(
+            send_ping=AsyncMock(side_effect=asyncio.TimeoutError()),
+            list_tools=AsyncMock(return_value=SimpleNamespace(tools=[])),
+        )
+
+        # Should NOT raise — the server is alive.
+        await task._keepalive_probe()
+
+        task.session.send_ping.assert_awaited_once()
+        task.session.list_tools.assert_awaited_once()
+        assert task._ping_unsupported is True
+
+    async def test_silent_ping_drop_both_fail_propagates(self):
+        """When both ping AND list_tools time out, it is a genuine liveness
+        failure — propagate so the caller reconnects."""
+        task = MCPServerTask("test")
+        task.initialize_result = _caps(tools=SimpleNamespace())
+        task.session = SimpleNamespace(
+            send_ping=AsyncMock(side_effect=asyncio.TimeoutError()),
+            list_tools=AsyncMock(side_effect=asyncio.TimeoutError()),
+        )
+
+        with pytest.raises((TimeoutError, asyncio.TimeoutError)):
+            await task._keepalive_probe()
+
+        assert task._ping_unsupported is False
+
+    async def test_silent_ping_drop_no_tools_propagates(self):
+        """A server that has no tools capability and times out on ping has no
+        fallback probe — the timeout must propagate immediately."""
+        task = MCPServerTask("test")
+        task.initialize_result = _caps(prompts=SimpleNamespace())  # no tools
+        task.session = SimpleNamespace(
+            send_ping=AsyncMock(side_effect=asyncio.TimeoutError()),
+            list_tools=AsyncMock(),
+        )
+
+        with pytest.raises((TimeoutError, asyncio.TimeoutError)):
+            await task._keepalive_probe()
+
+        # list_tools must not be called — no tools capability advertised.
+        task.session.list_tools.assert_not_called()
+        assert task._ping_unsupported is False
 
--- a/tools/mcp_tool.py
+++ b/tools/mcp_tool.py
@@ -2864,21 +2864,44 @@ class MCPServerTask:
                 await asyncio.wait_for(self.session.send_ping(), timeout=30.0)
                 return
             except Exception as exc:
-                # Only a "method not found" means ping is unsupported. Any
-                # other error (timeout, closed transport, session expired) is
-                # a real liveness failure — propagate so we reconnect.
-                if not _is_method_not_found_error(exc):
+                if _is_method_not_found_error(exc):
+                    # Structural -32601 or "Unknown method" — ping is
+                    # definitively unsupported.
+                    if not self._advertises_tools():
+                        raise
+                    self._ping_unsupported = True
+                    logger.info(
+                        "MCP server '%s': does not implement the optional "
+                        "'ping' utility (-32601); using 'list_tools' for "
+                        "keepalive on this connection.",
+                        self.name,
+                    )
+                elif isinstance(exc, (TimeoutError, asyncio.TimeoutError)) and self._advertises_tools():
+                    # A server that silently drops ping (no response at all)
+                    # produces a TimeoutError indistinguishable from a dead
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Distinguishes silent ping drop from unsupported transport (Closes #97245).

**What it does**
- `tools/mcp_tool.py:_keepalive_probe` now handles `TimeoutError` on `send_ping` specially: if `_is_method_not_found_error` → mark `_ping_unsupported` and fallback to `list_tools`; if timeout and `advertises_tools()` true, try `list_tools` — if succeeds, treat as alive and latch `_ping_unsupported=True` instead of reconnect-looping; else propagate timeout for genuine liveness failure.
- Tests cover `silent_ping_drop_falls_back_to_list_tools` (success → no raise, latch true), `both_fail_propagates`, and `no_tools_propagates` (no `list_tools` call when no tools capability).

**Non-blocking notes**
- Fallback uses `asyncio.TimeoutError` vs `TimeoutError` both — `pytest.raises((TimeoutError, asyncio.TimeoutError))` covers both.
- Ensure `list_tools` timeout is also 30s or shorter to not double wait; code path mirrors existing probe timeout.

Correct liveness classification fix.

Synxneuos:
Thanks for the review and confirmation @Enough1122!

Glad the liveness classification and fallback probe logic look solid. All unit tests and CI checks are **100% green** and passing. Ready for maintainer merge!

PR #95748, opened 2026-08-26
fix(install): put managed Node behind shims, stop registering it on PATH

for13to1:
# fix(install): put managed Node behind shims, stop registering it on PATH

## What does this PR do?

On Windows, `install.ps1` registers the Hermes-managed Node directory
(`%LOCALAPPDATA%\hermes\node`) itself as an entry in the **persisted User
PATH**, force-moving it to the front on every install and update
(`Set-ManagedNodeFirstOnUserPath`, move-to-front per #76459). That silently
overrode a user's own node/npm selection in every new shell, re-applied the
override after any manual reorder, and still lost to a system Node.js MSI in
Machine PATH (Windows combines PATH Machine-first + User-appended, so the
front-of-User-PATH placement is beatable). Hermes internals no longer depend
on this ordering — since #76459/#58150 every Node/npm call site resolves the
managed tree programmatically via `find_node_executable()` /
`iter_hermes_node_dirs()` — so only interactive shells were affected.

This change exposes node/npm/npx through **absolute-path `.cmd` delegators**
in `$HermesHome\bin` (the directory the `hermes.exe` launcher already lives
in, already on the persisted User PATH), stops registering the `\node`
directory itself, migrates existing installs by stripping the legacy entry,
and honors `HERMES_NODE_SKIP_LINKS=1` so set-up can provision the managed
tree privately with no shims and no PATH change — the same guard the POSIX
installer and health path already use. Python is unaffected throughout
(`venv\Scripts` is never on the persisted User PATH).

The behavior model after this PR, for a machine without Hermes:

1. System already has a suitable node ecosystem (node ≥ 26 + working npm)
   → Hermes uses it directly; installs nothing; touches no PATH. (Unchanged —
   and now *guaranteed*: see the shim-writer guards below.)
2. System has no suitable node → Hermes provisions its own managed tree under
   `%LOCALAPPDATA%\hermes\node`, reachable by absolute path.
3. Auto-management doesn't take over command names →
   `HERMES_NODE_SKIP_LINKS=1` makes set-up private-only — including when the
   flag is set *after* a non-private install (stale shims are removed).

## Related Issue

Fixes #95747

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `scripts/install.ps1`:
  - Removed `Set-ManagedNodeFirstOnUserPath` (the whole-directory-in-User-PATH
    move-to-front writer).
[...]

PR #97260, opened 2026-08-28
fix(uv): prefer the user's uv, move the managed one off PATH

### LINKED ISSUE #97259: [Bug]: Windows installer shadows the user's own uv with Hermes' managed uv
## Bug Description

On Windows, the installer places Hermes' managed uv binary at
`%LOCALAPPDATA%\hermes\bin\uv.exe` — and `%LOCALAPPDATA%\hermes\bin` is a
persisted User PATH entry (registered for the `hermes` launcher, prepended to
the front on a fresh install). The managed `uv.exe` therefore resolves as the
`uv` command in every new shell, silently shadowing a user's own uv install
(winget / scoop / `~\.local\bin` / `~\.cargo\bin`), even when that uv is
newer or configured the way the user wants.

This is the same bug class as #95747 (managed Node hijacking `node`/`npm`),
but for uv it is **worse in one respect**: the managed uv shares the `bin`
directory with `hermes.exe`, so the user cannot even remove the directory from
PATH without breaking the `hermes` command itself.

**Scope note:** POSIX is unaffected — `install.sh` only adds the `hermes`
command-link directory (`~/.local/bin` etc.) to PATH and never adds
`$HERMES_HOME/bin`, so the managed uv there stays private and the user's own
uv resolves normally. The asymmetry is Windows-only.

## Steps to Reproduce

1. Fresh-ish Windows box; install uv your own way (e.g.
   `winget install astral-sh.uv`, or the official installer, or cargo).
2. Run the canonical one-liner:
   `iex (irm https://hermes-agent.nousresearch.com/install.ps1)`
3. Open a new terminal, run `where uv`.
4. Observe `%LOCALAPPDATA%\hermes\bin\uv.exe` first — Hermes' managed binary,
   not yours.
5. Run `hermes update` (managed_uv.py self-updates the managed uv) — the `uv`
   version visible in your shells changes without you asking.

## Expected Behavior

Tiered, the same contract the managed-Node fix (#95748) establishes:

1. System already has a working uv → Hermes uses it directly; installs
   nothing; touches no PATH; never modifies the user's uv (no `uv self
   update` on a toolchain Hermes does not own). **Tier 1 wins even when a
   managed uv already exists** — the user's uv is the active one whenever
   present; the managed uv only takes over when the user's is gone (and may
   then self-heal via self-update).
2. System has no working uv → Hermes provisions its own managed uv under
   `%LOCALAPPDATA%\hermes\uv\`, reachable only by absolute path, never
   registered on PATH.
3. Auto-management never takes over command names → the managed uv is never
   the `uv` users resolve in their shells.

## Actual Behavior

- `uv` resolves to Hermes' managed binary in every new shell, even when the
  user has their own (possibly newer or differently configured) uv.
- The managed uv sits in the same directory as `hermes.exe`, so there is no
  PATH-level escape short of breaking the `hermes` command.
- `hermes update` silently changes the shell-visible `uv` version (managed_uv
  self-update).
- There is no guard equivalent to the POSIX `HERMES_NODE_SKIP_LINKS` concept
  — no way to request private-only uv provisioning.

## Affected Component

- [x] Setup / Installation

## Debug Report

_Pending — uploaded via `hermes debug share`; links will be pasted here._
(If the upload fails, provide `hermes debug share --local` output.)

## Operating System

Windows 11 (25H2)

## Python Version

3.11.x

## Hermes Version

v2026.8.19 (Hermes Agent v0.20.5) — latest release at time of reporting.

## Root Cause Analysis

`install.ps1` deliberately installs the managed uv with "no PATH probing, no
multi-location resolution chains" (`Install-Uv`), and `Set-PathVariable`
registers `$HermesHome\bin` (where `uv.exe` lives) on the persisted User
PATH. On POSIX, `install.sh` mirrors the managed-uv location
(`$HERMES_HOME/bin/uv`) but **never** adds `$HERMES_HOME/bin` to PATH — only
the `hermes` command-link directory is added — so POSIX shells resolve the
user's own uv. Windows has no such separation.

Hermes' internal call sites do not depend on PATH: `managed_uv_path()` in
`hermes_cli/managed_uv.py` resolves the managed binary by absolute path, and
`install.ps1`'s `Resolve-UvCmd` probes the managed location first. Only
interactive shells are affected — exactly the same shape as the pre-#95748
Node situation.

There is also no version floor for uv anywhere (no `uv --version` check
against a minimum) — `managed_uv.py` treats "runs" as sufficient and
self-updates to latest. That makes the "is the user's uv suitable?" tier-1
probe simple: it runs → use it.

## Proposed Fix

Design contract — three axes, calibrated separately:

- **Availability (which binary runs): coupled to the user.** Tier-1 first: a
  usable `uv` on PATH (runs `uv --version`) wins, even when a managed uv
  already exists. The managed binary is only provisioned when the user has no
  working uv — and only then.
- **Ownership (who may mutate it): hard-decoupled — the red line.** The
  user's uv is used read-only: never `uv self update`, never moved. The
  self-update path (`update_managed_uv`) already resolves through
  managed-only `resolve_uv()`, so it can never touch a toolchain Hermes does
  not own.
- **Artifacts (venv/python construction): config isolated, engine
  inherited.** Hermes' venv operations keep pinning `UV_PYTHON_INSTALL_DIR`
  to Hermes' own python dir and neutralizing user uv config — via the
  existing `managed_python_env()` — regardless of which engine runs them.

Implementation (single resolution seam):

- `hermes_cli/managed_uv.py`: `resolve_uv_engine()` returns
  `{engine: user|managed|none, path, can_self_update}`; `ensure_uv()` uses it
  tier-1 first; `managed_uv_path()` moves to `$HERMES_HOME/uv/uv(.exe)` — a
  private directory never registered on PATH; legacy `$HERMES_HOME/bin/uv`
  is migrated once (best-effort `os.replace`).
- `scripts/install.ps1` / `scripts/install.sh`: `Install-Uv`/`install_uv()`
  probe the user's uv first, install the managed binary into the private dir
  (`UV_INSTALL_DIR`/`UV_UNMANAGED_INSTALL` retargeted), and migrate a legacy
  `bin\uv.exe` / `bin/uv`.
- `hermes_cli/_early_recovery.py`: managed path updated (legacy path kept as
  a fallback candidate).
- Tests: Python resolver/migration/tier-1 coverage (temp HERMES_HOME) +
  PowerShell AST-lift tests for the install.ps1 helpers.

PR is ready: see linked PR.

Related: #95747 / PR #95748 fix the same class for Node; this issue is the
uv instance.

for13to1:
# fix(uv): prefer the user's uv, move the managed one off PATH

## What does this PR do?

On Windows, `install.ps1` places Hermes' managed uv at
`%LOCALAPPDATA%\hermes\bin\uv.exe` — and `%LOCALAPPDATA%\hermes\bin` is a
persisted User PATH entry (registered for the `hermes` launcher, prepended on
fresh install). The managed `uv.exe` therefore resolves as the `uv` command in
every new shell, silently shadowing the user's own uv — and because it shares
`bin` with `hermes.exe`, the user cannot even remove the directory from PATH
without breaking the `hermes` command. `hermes update` then silently changes
the shell-visible `uv` version (managed_uv self-update). POSIX is unaffected:
`install.sh` never adds `$HERMES_HOME/bin` to PATH, so the asymmetry is
Windows-only.

This change applies the same tiered contract the managed-Node installer
established (PR #95748):

1. **System already has a working uv → use it directly; install nothing;
   touch no PATH; never modify it** (no `uv self update` on a toolchain
   Hermes does not own). Tier-1 wins even when a managed uv already exists.
2. **No working uv → provision a managed uv under `$HERMES_HOME/uv/`** — a
   private directory that is never registered on PATH — reachable only by
   absolute path.
3. **Auto-management never takes over command names** — shells always resolve
   the user's own uv (or none), never Hermes' copy.

## Related Issue

Fixes #97259

## Type of Change

## Changes Made

- `hermes_cli/managed_uv.py`:
  - `managed_uv_path()` → `$HERMES_HOME/uv/uv` (POSIX) / `$HERMES_HOME\uv\uv.exe`
    (Windows) — the private location, never on PATH. `legacy_managed_uv_path()`
    recognizes the pre-isolation `bin/uv(.exe)` shape.
  - New `resolve_uv_engine()` — the single resolution seam. Tier-1: a `uv` on
    PATH that passes the `--version` smoke test is adopted as `user`
    (`can_self_update=False`); a PATH hit that resolves to Hermes' own managed
    or legacy binary is never mistaken for the user's (Windows
    case-insensitive comparison). Then the managed binary; then `none`.
  - `ensure_uv()` / `_ensure_uv_path()` — tier-1 first: a usable user uv
    returns without any managed provisioning; otherwise the managed binary is
    installed into the private dir, migrating a legacy `bin/uv(.exe)` first
    (best-effort `os.replace`, no reinstall after a successful move).
  - `update_managed_uv()` — self-update stays managed-only (`resolve_uv()`
    is the target resolver), but the CVE-driven runtime-repair probe now runs
    with ANY usable uv (user's or managed): with tier-1 there is often no
    managed binary at all, and repair must not be gated behind its existence.
    The probe/provisioning writes stay in Hermes' own dirs via
    `managed_python_env()`, and `_refresh_managed_uv_catalog()` refuses to
    touch a foreign binary — the user's uv is used read-only, never modified.
- `scripts/install.ps1`:
  - New helpers `Get-ManagedUvPath`, `Test-UserUvUsable` (tier-1 probe with
    the managed/legacy guard), `Move-LegacyManagedUv` (one-time migration),
    `Set-UvPythonIsolationEnv` (pins `UV_PYTHON_INSTALL_DIR` to
    `$HermesHome\python` + `UV_PYTHON_INSTALL_BIN=0` +
    `UV_PYTHON_INSTALL_REGISTRY=0`, wired into `Resolve-UvCmd` so every uv
    python write stays inside Hermes' tree regardless of which uv engine runs).
  - `Install-Uv` — tier-1 first; managed binary to the private dir
    (`UV_INSTALL_DIR` retargeted); legacy migration on upgrade.
  - `Resolve-UvCmd` — tier-1 first, then the private managed path, then
    migration, then the existing PATH fallbacks.
- `scripts/install.sh`:
  - `install_uv()` — tier-1 (`command -v uv` + runs) first; managed binary to
    `$HERMES_HOME/uv/uv`; legacy `bin/uv` migration; `UV_UNMANAGED_INSTALL`
    retargeted; non-root installs pin `UV_PYTHON_INSTALL_DIR` to
    `$HERMES_HOME/python` (root FHS keeps its world-readable pin).
- `hermes_cli/_early_recovery.py` — managed path updated; legacy `bin/uv`
  kept as a fallback candidate.
- `hermes_cli/update_cmd.py` — docstring path reference updated.
- `website/docs/user-guide/windows-native.md` + zh-Hans — `bin\` is now
  documented as launchers-only; the private `uv\` directory added.

### Scope note (why the diff touches consumers beyond the installer)

Moving the managed uv out of `$HERMES_HOME/bin` into the private
`$HERMES_HOME/uv` dir has a downstream: several runtime consumers still
resolved the managed binary at its old home. `tools/browser_use_cli.py`'s
uvx zero-install probe and `tools/environments/local.py`'s terminal-subshell
PATH injection would silently lose the managed uv after this change. Those
files are patched here for consistency (`$HERMES_HOME/uv` probed/appended,
gap-fill only — a user's own uv on PATH still wins, first-occurrence-wins),
with stale `$HERMES_HOME/bin` mention synced in comments/docs. Without this,
the fix itself would be the thing that breaks uv discovery.

## How to Test

1. **Python** (cross-platform, no network):
   `scripts/run_tests.sh tests/hermes_cli/test_managed_uv.py tests/test_managed_runtime_resolution.py`
   — covers the resolver (user/managed/none), the managed/legacy guard, the
   legacy migration, and tier-1 `ensure_uv` (user uv → no managed install;
   broken user uv → managed fallback).
2. **PowerShell** (any machine with pwsh):
   `pwsh -NoProfile -File scripts/ci/test_install_ps1_uv_isolation.ps1`
   — AST-lifts the real `Get-ManagedUvPath` / `Test-UserUvUsable` /
   `Move-LegacyManagedUv` / `Set-UvPythonIsolationEnv` bodies and asserts the
   private path, tier-1 guard, migration semantics, and write containment
   (inherited `UV_PYTHON_INSTALL_DIR` is overridden, never respected).
3. **Windows E2E** (needs a Windows box): install with a user-scoped uv of
   your own → `where uv` resolves to yours, not `%LOCALAPPDATA%\hermes\...`;
   re-run the installer with no user uv → managed uv lands in
   `%LOCALAPPDATA%\hermes\uv\` and `where uv` finds nothing Hermes-owned;
   upgrade an install that has the old `bin\uv.exe` → it migrates to `uv\`.

### Code

- [x] I've run the relevant tests for this change:
  - `tests/hermes_cli/test_managed_uv.py` → 53 passed (incl. tier-1 repair
    regression: the repair probe runs with the user's uv when no managed
    binary exists)
  - `tests/test_managed_runtime_resolution.py` → 7 passed
  - `pwsh -NoProfile -File scripts/ci/test_install_ps1_uv_isolation.ps1` → all
    assertions passed; `install.ps1` / `install.sh` parse clean
- [x] I've added tests for my changes (required for bug fixes)
- [x] I've tested on my platform: macOS (Python + PowerShell harness); Windows
  E2E pending (no Windows host available)

- [ ] I've updated relevant documentation — N/A (installer/runtime behavior;
  no user-facing config key added)
- [x] I've considered cross-platform impact (Windows, macOS) per the
  compatibility guide — the managed uv is now private on BOTH platforms; the
  Windows-specific PATH shadowing is the bug being fixed
- [ ] I've updated tool descriptions/schemas — N/A (no model tool changed)

## Known Limitations

- The tier-1 suitability check is "runs `uv --version`" — there is still no
  uv version floor (the codebase never had one; the managed uv self-updates).
  If a floor is ever needed, `resolve_uv_engine()` is the single place to add
  it.
- The installer's Rung-3 salvage (copying an existing uv into the managed
  location when the network is fully blocked) still targets the private dir;
  it only runs when no usable user uv exists.

## Screenshots / Logs

```
$ where uv            # after install, on a box with the user's own uv
C:\Users\me\AppData\Local\Programs\uv\uv.exe     (the user's — not Hermes')

$ ls %LOCALAPPDATA%\hermes\uv\
uv.exe                                             (Hermes' private copy, off PATH)
```

`pwsh -NoProfile -File scripts/ci/test_install_ps1_uv_isolation.ps1`
→ `all assertions passed`.

### DIFF
--- a/hermes_cli/_early_recovery.py
+++ b/hermes_cli/_early_recovery.py
@@ -23,7 +23,6 @@ in main.py, which runs right after import succeeds.
 
 from __future__ import annotations
 
-import importlib
 import os
 import shutil
 import subprocess
@@ -281,7 +280,7 @@ def _probe_broken_packages() -> list[str]:
     broken: list[str] = []
     for mod_name, attr in LAZY_REFRESH_IMPORT_PROBES:
         try:
-            mod = importlib.import_module(mod_name)
+            mod = __import__(mod_name)
             if not hasattr(mod, attr):
                 raise ImportError(f"{mod_name} missing {attr}")
             if mod_name == "certifi" and _certifi_bundle_broken():
@@ -302,10 +301,22 @@ def _find_uv_binary() -> str | None:
     (``~/.hermes/uv/uv.exe``) or the user has on PATH.  Stdlib-only.
     """
     exe = "uv.exe" if sys.platform == "win32" else "uv"
+
+    hermes_env = os.environ.get("HERMES_HOME", "").strip()
+    if hermes_env:
+        hermes_home = Path(hermes_env)
+    elif sys.platform == "win32":
+        local_appdata = os.environ.get("LOCALAPPDATA", "").strip()
+        base = Path(local_appdata) if local_appdata else Path.home() / "AppData" / "Local"
+        hermes_home = base / "hermes"
+    else:
+        hermes_home = Path.home() / ".hermes"
+
     candidates = [
-        Path.home() / ".hermes" / "uv" / exe,
+        # Private uv in Hermes home (custom HERMES_HOME, %LOCALAPPDATA%/hermes on Windows, ~/.hermes on POSIX)
+        hermes_home / "uv" / exe,
         # Legacy pre-isolation layout; migrated to the private dir on use.
-        Path.home() / ".hermes" / "bin" / exe,
+        hermes_home / "bin" / exe,
         Path.home() / ".local" / "bin" / exe,
         Path.home() / ".cargo" / "bin" / exe,
     ]
--- a/hermes_cli/_early_recovery.py
+++ b/hermes_cli/_early_recovery.py
@@ -299,10 +299,12 @@ def _find_uv_binary() -> str | None:
     uv-managed base interpreters carry an ``EXTERNALLY-MANAGED`` marker, so
     the stdlib ``pip`` fallback below refuses to touch them.  In that state
     the only sanctioned installer is uv itself, which Hermes already vendors
-    (``~/.hermes/bin/uv.exe``) or the user has on PATH.  Stdlib-only.
+    (``~/.hermes/uv/uv.exe``) or the user has on PATH.  Stdlib-only.
     """
     exe = "uv.exe" if sys.platform == "win32" else "uv"
     candidates = [
+        Path.home() / ".hermes" / "uv" / exe,
+        # Legacy pre-isolation layout; migrated to the private dir on use.
         Path.home() / ".hermes" / "bin" / exe,
         Path.home() / ".local" / "bin" / exe,
         Path.home() / ".cargo" / "bin" / exe,
--- a/hermes_cli/_install_repair.py
+++ b/hermes_cli/_install_repair.py
@@ -190,9 +190,10 @@ def ensure_windows_bin_launchers(
     On Windows, ``hermes`` resolves through launchers derived from the venv
     console scripts — never ``venv\\Scripts`` itself on PATH, which would
     shadow the user's ``python`` (#83797). The canonical launcher home is
-    the managed binary dir — the default Hermes root's ``bin``
-    (``%LOCALAPPDATA%\\hermes\\bin``, next to the managed uv) — which lives
-    OUTSIDE the git checkout so no git operation can ever touch it. It is
+    the default Hermes root's ``bin`` (``%LOCALAPPDATA%\\hermes\\bin`` — the
+    managed uv lives in the sibling ``uv\\`` dir since the uv-isolation
+    change) which lives OUTSIDE the git checkout so no git operation can
+    ever touch it. It is
     a per-machine dir shared by every profile: ``get_hermes_home()`` would
     point inside ``profiles\\<name>`` under ``hermes -p``, so the anchor
     here is :func:`hermes_constants.get_default_hermes_root`.
--- a/hermes_cli/managed_uv.py
+++ b/hermes_cli/managed_uv.py
@@ -1598,8 +1598,8 @@ def _install_uv(target: Path) -> None:
     """Bootstrap uv into *target* using the official standalone installer.
 
     Uses ``UV_UNMANAGED_INSTALL`` (POSIX) or ``UV_INSTALL_DIR`` (Windows)
-    so the astral installer writes the binary directly into
-    ``$HERMES_HOME/bin/`` instead of ``~/.local/bin/``.
+    so the astral installer writes the binary directly into the private
+    managed dir (``$HERMES_HOME/uv``) instead of ``~/.local/bin/``.
     """
     system = platform.system()
     env = {
--- a/hermes_cli/managed_uv.py
+++ b/hermes_cli/managed_uv.py
@@ -112,6 +112,11 @@ def _uv_runs(path: str) -> bool:
     *depends* on one: a uv that is on PATH yet cannot even report its version
     (broken install, wrong arch, quarantined binary) is not "suitable" and
     falls through to the managed fallback.
+
+    Worst case is a bounded 15s: a *hung* PATH uv (network-sandboxed or
+    quarantined binary that never answers) stalls resolution for 15s before
+    falling through.  Resolution runs only at update/install/bootstrap time
+    — never on a per-turn hot path — so the bound is deliberate.
     """
     try:
         result = subprocess.run(
@@ -124,6 +129,23 @@ def _uv_runs(path: str) -> bool:
     except (OSError, subprocess.SubprocessError):
         return False
     return result.returncode == 0
+
+
+def _is_managed_uv(uv_bin: str | Path) -> bool:
+    """True when *uv_bin* is Hermes' own private managed uv binary.
+
+    Comparison is case-insensitive on Windows (``os.path.normcase`` lowercases
+    there and is a no-op on POSIX) — the same rule :func:`resolve_uv_engine`
+    applies so a PATH hit that IS our binary is never mistaken for the user's.
+    This is the ownership gate: engine-mutating operations (catalog refresh,
+    self-update) may only run against the managed binary, never the user's.
+    """
+    try:
+        return os.path.normcase(str(Path(uv_bin).resolve())) == os.path.normcase(
+            str(managed_uv_path().resolve())
+        )
+    except OSError:
+        return False
 
 
 @dataclass
@@ -1306,15 +1328,21 @@ def _refresh_managed_uv_catalog(uv_bin: str) -> bool:
     provisioning retry can now see a different catalog.  ``False`` means a
     retry would resolve identically and is not worth the download cycle.
     """
-    managed = managed_uv_path()
-    try:
-        if Path(uv_bin).resolve() != managed.resolve():
-            return False
-    except OSError:
+    # Ownership red line: this re-bootstraps the engine (the only supported
+    # refresh path for UV_UNMANAGED_INSTALL binaries), so it may only ever
+    # run against Hermes' own managed binary.  A foreign engine — typically
+    # the user's own uv under tier-1 — is used strictly read-only here and
+    # never re-bootstrapped.
+    if not _is_managed_uv(uv_bin):
+        logger.debug(
+            "refusing to refresh uv catalog for foreign binary %s "
+            "(read-only: the user's uv is never re-bootstrapped)",
+            uv_bin,
+        )
         return False
     before = _uv_version_string(uv_bin)
     try:
-        _install_uv(managed)
+        _install_uv(managed_uv_path())
     except Exception as exc:
         logger.warning("managed uv refresh failed: %s", exc)
         return False
@@ -1402,6 +1430,22 @@ def repair_vulnerable_runtime(
     root = Path(project_root) if project_root is not None else _PROJECT_ROOT
     live = Path(venv_dir) if venv_dir is not None else _default_live_venv(root)
     live_python = _venv_python(live)
+
+    # Ownership red line: the repair may run with the USER's own uv as the
+    # engine (tier-1 — there is often no managed binary at all), but that
+    # engine is strictly read-only.  Every uv invocation below writes only
+    # into Hermes' own directories: managed_python_env() pins the python
+    # install dir, shims and registry, and the venv/sync target the
+    # checkout-scoped candidate.  The only engine-mutating operation in this
+    # path (_refresh_managed_uv_catalog) refuses a foreign binary.  A future
+    # edit that adds any engine-mutating call (``uv self update``,
+    # ``uv tool install``, ...) MUST gate it on _is_managed_uv(uv_bin) first.
+    if not _is_managed_uv(uv_bin):
+        logger.info(
+            "runtime repair engine is the user's uv (%s) — used strictly "
+            "read-only; all writes stay inside Hermes' own directories",
+            uv_bin,
+        )
     if not (root / "pyproject.toml").is_file() or not live_python.is_file():
         return RuntimeRepairResult("not-applicable")
 
--- a/hermes_cli/managed_uv.py
+++ b/hermes_cli/managed_uv.py
@@ -1,11 +1,21 @@
 """Hermes-managed uv and Python runtime repair.
 
-Hermes owns its own uv binary at ``$HERMES_HOME/bin/uv`` (or ``uv.exe`` on
-Windows).  Every code path that needs uv resolves it from that single location.
-If the binary is missing, ``ensure_uv()`` bootstraps it via the official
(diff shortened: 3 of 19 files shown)


Enough1122:

Clean, well-tested change. Moving the managed uv to a private `$HERMES_HOME/uv` (never on PATH), preferring the user's own uv read-only (tier-1), and adding the one-time legacy migration are all coherent, and the "ownership red line" (`update_managed_uv` self-updates only `resolve_uv()`'s managed binary) is the right invariant to preserve. The `normcase` comparison for Windows case-insensitivity in `resolve_uv_engine()` is correct, and the Python tests plus the AST-lift PowerShell harness cover the important branches (user-uv wins, broken user-uv falls through, managed/legacy PATH hits are not "user", migration moves/reuses without reinstall).

A few points to consider:

1. **install.sh vs install.ps1 disagree on `UV_PYTHON_INSTALL_DIR` inheritance.** `Set-UvPythonIsolationEnv` (install.ps1) unconditionally overwrites the inherited `UV_PYTHON_INSTALL_DIR` — and the new `test_install_ps1_uv_isolation.ps1` explicitly asserts that override. But install.sh only sets it when unset (`if [ -z "${UV_PYTHON_INSTALL_DIR:-}" ]`), i.e. it *respects* an inherited value. The comment in both says "Overrides any inherited values", so one of them is wrong relative to the stated intent. Either make both override (matching the test) or both respect the user's value — and pick the documented contract. Also `UV_PYTHON_INSTALL_BIN`/`_REGISTRY` are unconditional in both, so only DIR differs; that asymmetry looks unintentional.

2. **`repair_vulnerable_runtime(repair_uv)` runs against the *user's* uv when no managed binary exists — the read-only guarantee deserves an explicit guard, not just a comment.** The docstring asserts "Read-only use of the user's engine — never modified", and `test_repair_runs_with_user_uv_when_no_managed_exists` pins that the probe is invoked with the user's uv path. That's only safe if `repair_vulnerable_runtime` (and anything it calls, e.g. `_refresh_managed_uv_catalog`) never writes to or self-updates the passed engine. Since the whole point of tier-1 is "never touch the user's toolchain", consider asserting inside the repair path that a `"user"` engine is treated as strictly read-only (or refuse to run repair with a user engine and fall back to "none"), so a future edit can't accidentally mutate the user's uv.

3. **`_early_recovery.py` adds `~/.hermes/uv/uv.exe` hardcoded via `Path.home()`, not `get_hermes_home()`.** It follows the pre-existing `~/.hermes/bin` pattern in that module (which is stdlib-only by design), so it's consistent — but worth noting it means profile-scoped installs (`HERMES_HOME` pointing elsewhere) won't find the private uv here; only the default profile does. If `_early_recovery` can't import `hermes_constants`, at least leave a comment documenting that this is default-profile-only, since the legacy `bin` candidate has the same limitation.

4. Minor: `_uv_runs` uses a 15s timeout, so a *hung* PATH uv (network-sandboxed or quarantined binary that never answers `--version`) stalls every `resolve_uv_engine()` call for 15s before falling through to managed. Acceptable as a bounded worst case, but if this ever runs on a hot path it's worth remembering.

Nice work on the test coverage — particularly the PowerShell harness lifting real function bodies rather than reimplementing the logic.

for13to1:
> > AI code review — automated review for reference, author can ignore or act on any point.
> 
> Clean, well-tested change. Moving the managed uv to a private `$HERMES_HOME/uv` (never on PATH), preferring the user's own uv read-only (tier-1), and adding the one-time legacy migration are all coherent, and the "ownership red line" (`update_managed_uv` self-updates only `resolve_uv()`'s managed binary) is the right invariant to preserve. The `normcase` comparison for Windows case-insensitivity in `resolve_uv_engine()` is correct, and the Python tests plus the AST-lift PowerShell harness cover the important branches (user-uv wins, broken user-uv falls through, managed/legacy PATH hits are not "user", migration moves/reuses without reinstall).
> 
> A few points to consider:
> 
> 1. **install.sh vs install.ps1 disagree on `UV_PYTHON_INSTALL_DIR` inheritance.** `Set-UvPythonIsolationEnv` (install.ps1) unconditionally overwrites the inherited `UV_PYTHON_INSTALL_DIR` — and the new `test_install_ps1_uv_isolation.ps1` explicitly asserts that override. But install.sh only sets it when unset (`if [ -z "${UV_PYTHON_INSTALL_DIR:-}" ]`), i.e. it _respects_ an inherited value. The comment in both says "Overrides any inherited values", so one of them is wrong relative to the stated intent. Either make both override (matching the test) or both respect the user's value — and pick the documented contract. Also `UV_PYTHON_INSTALL_BIN`/`_REGISTRY` are unconditional in both, so only DIR differs; that asymmetry looks unintentional.
> 2. **`repair_vulnerable_runtime(repair_uv)` runs against the _user's_ uv when no managed binary exists — the read-only guarantee deserves an explicit guard, not just a comment.** The docstring asserts "Read-only use of the user's engine — never modified", and `test_repair_runs_with_user_uv_when_no_managed_exists` pins that the probe is invoked with the user's uv path. That's only safe if `repair_vulnerable_runtime` (and anything it calls, e.g. `_refresh_managed_uv_catalog`) never writes to or self-updates the passed engine. Since the whole point of tier-1 is "never touch the user's toolchain", consider asserting inside the repair path that a `"user"` engine is treated as strictly read-only (or refuse to run repair with a user engine and fall back to "none"), so a future edit can't accidentally mutate the user's uv.
> 3. **`_early_recovery.py` adds `~/.hermes/uv/uv.exe` hardcoded via `Path.home()`, not `get_hermes_home()`.** It follows the pre-existing `~/.hermes/bin` pattern in that module (which is stdlib-only by design), so it's consistent — but worth noting it means profile-scoped installs (`HERMES_HOME` pointing elsewhere) won't find the private uv here; only the default profile does. If `_early_recovery` can't import `hermes_constants`, at least leave a comment documenting that this is default-profile-only, since the legacy `bin` candidate has the same limitation.
> 4. Minor: `_uv_runs` uses a 15s timeout, so a _hung_ PATH uv (network-sandboxed or quarantined binary that never answers `--version`) stalls every `resolve_uv_engine()` call for 15s before falling through to managed. Acceptable as a bounded worst case, but if this ever runs on a hot path it's worth remembering.
> 
> Nice work on the test coverage — particularly the PowerShell harness lifting real function bodies rather than reimplementing the logic.

Thanks! Addressed in three follow-ups:

1fcd37d5 — review points 1–4: install.sh now overrides inherited UV_PYTHON_INSTALL_DIR like install.ps1; _is_managed_uv() ownership gate + read-only guard in the repair path; _early_recovery.py lookup is now
 HERMES_HOME/LOCALAPPDATA-aware; documented the 15s _uv_runs bound.

911a32e8 — synced consumers to the new $HERMES_HOME/uv location (browser-use uvx probe, terminal subshell PATH, stale bin/ comments).

823b29f9 — _install_repair.py docstring fix.

Tests added for the read-only contract + foreign-binary refusal; suites green.

PR #97261, opened 2026-08-28
fix(code-execution): preserve remote kernel output contract

### LINKED ISSUE #97256: Remote session kernels drop exit_code and clipped stdout recovery metadata
## Summary

The persistent remote-kernel path does not preserve the `execute_code` result contract used by the per-call remote path and the local session kernel:

1. `_finish_remote_kernel_result()` omits `exit_code` entirely.
2. The remote runner clips stdout before returning it and sets `stdout_clipped=true`, but the formatter ignores that flag. Because the already-clipped text is at the host limit, the host reports `stdout_truncated=false` and zero omitted bytes.
3. #97043 added full-output spillover for per-call execution and local session kernels, but did not update `tools/code_kernel_remote.py`. The remote runner therefore discards the omitted output permanently.

## Current behavior

A remote-kernel result with clipped output produces:

```text
exit_code: missing
stdout_truncated: false
stdout_bytes_total: 50000
stdout_bytes_omitted: 0
```

The raw cell result already carries `stdout_clipped=true`, but `_finish_remote_kernel_result()` never consumes it.

## Expected behavior

Persistent remote kernels should match the other `execute_code` paths:

- always return an `exit_code`
- report runner-side clipping as truncation with faithful byte metadata
- spill the complete cell output in the remote kernel directory and return a `stdout_spill_path` plus the normal `read_file` recovery guidance

## Affected code

- `tools/code_kernel_remote.py`: runner-side clipping discards the full text
- `tools/code_execution_tool.py`: remote-kernel result formatter drops the clipping flags and `exit_code`

The remote-kernel implementation states that it preserves the same output post-processing invariants as the local kernel, so this is a parity regression rather than a new behavior request.

Liuzikaii:
## Summary

- preserve `exit_code` on persistent remote-kernel results
- report runner-side clipping with faithful truncation byte metadata
- spill complete remote cell stdout before clipping and return the remote path with `read_file` recovery guidance
- add a real subprocess/file-protocol regression test for runner spillover

Fixes #97256

## Testing

- `python -m pytest -p no:cacheprovider tests/tools/test_code_kernel_remote.py tests/tools/test_code_kernel.py -q`
- `python -m pytest -p no:cacheprovider tests/tools/test_code_execution.py -q -k 'not missing_token_rejected'`
- `python -m ruff check tools/code_kernel_remote.py tools/code_execution_tool.py tests/tools/test_code_kernel_remote.py`

The deselected test requires `socket.AF_UNIX`, which is unavailable in the native Windows test environment.

### DIFF
--- a/tests/tools/test_code_kernel_remote.py
+++ b/tests/tools/test_code_kernel_remote.py
@@ -10,7 +10,9 @@ state_lost/state_reset reporting, fail-open, and owner isolation.
 """
 import json
 import os
+import subprocess
 import sys
+import tempfile
 import unittest
 from unittest.mock import patch
 
@@ -18,6 +20,7 @@ sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", ".."))
 
 from tools.code_kernel_remote import (
     _REMOTE_KERNELS,
+    REMOTE_KERNEL_RUNNER_SOURCE,
     RemoteKernel,
     execute_in_remote_kernel,
     shutdown_all_remote_kernels,
@@ -117,6 +120,7 @@ class TestSpawnAndReuse(RemoteKernelBase):
         self.assertEqual(
             sum(1 for c in env.commands if "nohup" in c), 1,
         )
+        self.assertEqual(first["exit_code"], 0)
 
     def test_spawn_failure_fails_open(self):
         env = ScriptedEnv([
@@ -208,7 +212,7 @@ class TestDispatchIntegration(unittest.TestCase):
 
         fake = {
             "status": "success", "stdout": "kernel says hi\n", "stderr": "",
-            "traceback": "", "tool_calls_made": 0,
+            "traceback": "", "tool_calls_made": 0, "exit_code": 0,
             "kernel": {"reused": True, "remote": True, "execution_count": 3},
         }
         env = ScriptedEnv([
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Preserves remote kernel stdout spill / clipping contract.

**What it does**
- `tests/tools/test_code_kernel_remote.py` / `tools/code_kernel_remote.py` + `tools/code_execution_tool.py:_execute_remote` forward `exit_code:0`, `stdout_clipped`, `stdout_bytes_total`, `stdout_spill_path`, `stdout_bytes_omitted` and `stdout_truncated` mapping; remote runner (`REMOTE_KERNEL_RUNNER_SOURCE`) spills clipped stdout to `cells/cell_*.txt` before exit.
- Test `test_execute_remote_preserves_runner_clipping_metadata` asserts spill path, totals, and `warning` includes `read_file` truncation hint; `test_clipped_stdout_is_spilled_before_runner_exit` uses real kernel_runner with `capture_limit=100` to ensure file exists.

**Non-blocking notes**
- `exit_code` added to remote result — ensure `code_execution_tool` previously dropped it; now forwarded correctly.
- `capture_limit` / `spill_cap` formatting via `.format(capture_limit=100, spill_cap=1_000)` — verify spill_cap 1000 vs dynamic from `_load_config`; test uses fixed values fine.

Preserves local/remote output contract parity.

Issue #81995, opened 2026-08-08
[Bug]: Stalled stdio MCP cold-spawn leaves in-flight tool call attached to dead subprocess for full 300s timeout — no fail-fast on supervisor respawn


menhguin:
## Bug Description

When an idle-killed stdio MCP server is lazily respawned by a tool call, the fresh subprocess can stall **silently** (zero bytes on stdout/stderr — no startup output at all). Hermes's supervisor correctly detects the dead spawn and starts a replacement ~60–85s later, **but the in-flight tool call that triggered the first spawn is never failed fast or re-routed** — it stays attached to the dead subprocess until the full configured tool timeout (300s default) fires.

Net effect: a transient spawn stall turns a sub-second tool call into a 5-minute hang, even though a healthy server instance already exists by the ~1-minute mark.

Observed twice in one night on the same server (intermittent — 2 of 8 cold spawns stalled).

## Environment

- Hermes Agent v0.20.0 (2026.8.3), brew install, Python 3.14
- macOS 26.6 (Apple Silicon)
- Server config: stdio-over-SSH wrapper — `ssh -T -o BatchMode=yes <host> /opt/homebrew/bin/npx -y chrome-devtools-mcp@latest --autoConnect --no-usage-statistics`

## Steps to Reproduce

1. Configure a stdio MCP server (in our case wrapped in `ssh`, which may make the stall more likely — a slow remote process start produces no local output).
2. Let it idle past the ~5-minute idle-kill so the next call triggers a cold spawn.
3. Call any tool on that server.
4. If the spawn stalls (intermittent), observe: the tool call hangs the full 300s.

Log evidence from `~/.hermes/logs/mcp-stderr.log` (timestamps, two independent occurrences):

```
===== [03:24:34] starting MCP server 'chrome-devtools-behemoth' =====   ← stalled spawn: ZERO further output (a healthy spawn prints the server's startup banner within ~1s)
===== [03:25:39] starting MCP server 'chrome-devtools-behemoth' =====   ← supervisor respawn 65s later; starts cleanly, prints banner, serves requests
# ...meanwhile the original tool call is still waiting on the 03:24:34 instance...
# 03:29:34 — tool call returns: {"error": "MCP call failed: TimeoutError: MCP call timed out after 300.1s (configured timeout: 300.0)"}
```

[...]

PR #97262, opened 2026-08-28
fix: inverted liveness check fast-failed healthy stdio MCP servers

Fyuturevizion:
## Summary
- `MCPServerTask._stdio_children_dead()` documented "True when every stdio child has exited" but the alive branch returned `True`, so **every** stdio MCP tool call against a healthy server was cancelled immediately with "MCP stdio subprocess for 'X' has exited".
- Return `False` when any tracked child is alive; `True` only when every tracked PID has exited, matching the docstring.

## Reproduction
Any stdio MCP server with captured child PIDs (e.g. `uvx blender-mcp`). Every tool call fast-fails while the child process runs fine:
```
MCP stdio subprocess for 'blender' has exited
```

## Verification
- One-shot agent session end-to-end through two stdio MCP servers (blender + filesystem) after the fix: 25-tool server connects, tool calls execute, full Cycles render pipeline completed.
- Blender MCP socket log confirms tool commands now reach the server (previously zero commands arrived). The render artifact exists on disk with expected dimensions/size (1280x720 PNG, ~887KB).

### DIFF
--- a/tools/mcp_tool.py
+++ b/tools/mcp_tool.py
@@ -2994,9 +2994,8 @@ class MCPServerTask:
 
             if not psutil.pid_exists(pid):
                 continue  # this one is dead
-            return True  # alive (signal permission irrelevant for liveness)
-            return False  # at least one child alive
-        return True
+            return False  # at least one child alive -> not dead
+        return True  # every tracked child exited
 
     async def _watch_stdio_children(self) -> None:
         """Poll child liveness while a stdio RPC is in flight (#81995).


Enough1122:

**Overall:** One-line liveness fix — inverted stdio child check (#81995).

**What it does**
- `tools/mcp_tool.py:_stdio_children_dead` had `return True` / `return False` swapped and dead code `return True` unreachable after `return False` (copy-paste). Now `if pid_exists → return False` (not dead, alive), final `return True` (every child exited).

**Non-blocking notes**
- Previously `psutil.pid_exists` failure returned `True` (dead) vs current dead-check — but inverted logic fast-failed healthy servers; fix restores correct keepalive.

Trivial, correct.

kshitijk4poor:
Closing as already fixed on `main` — this exact inversion was corrected in 2663117f72 ("fix(mcp): correct inverted liveness check in _stdio_children_dead"), with a follow-up in 8aae2ea539 that also signals reconnect from the mid-call fast-fail site plus regression tests. The landed predicate matches this PR's logic (return False on the first alive child; True only when all tracked children exited; probe failures treated as unknown → don't fail fast).

Your diagnosis was spot-on — it just raced the fix that was already in flight. Thanks!

PR #94647, opened 2026-08-25
feat(tools): session-persistent kernels for execute_code (kernel_mode: session)

rlaope:
## What does this PR do?

`execute_code` spawns a fresh Python process per call, so every multi-step data task re-loads its inputs: a CSV parsed in call one is gone by call two, and scripts end up routing state through temp files to survive. Hermes already rewards programmatic tool calling — execute_code-only turns refund the iteration budget (`agent/conversation_loop.py`) — which makes the missing half, state that survives between calls, the practical bottleneck.

This adds an opt-in `code_execution.kernel_mode: session`: one persistent kernel per (task, mode, interpreter, cwd, tool-set). Variables, imports, and loaded data persist across calls; `reset=true` discards state on demand. The default `per-call` keeps today's behavior byte-for-byte.

**Safety posture is unchanged by design:**
- The child environment comes from the same builder as the per-call path — extracted into `_build_child_env()` rather than duplicated, so the secret scrubbing / UTF-8 / TZ / PYTHONPATH-hygiene rules cannot drift between the two paths.
- The RPC server is the same `_rpc_server_loop` with the same per-kernel token; the tool-call budget applies **per cell** (counter resets each call).
- Output passes the same ANSI strip + secret redaction as per-call.
- A timed-out or interrupted cell kills the whole kernel process tree and the next call respawns fresh — a wedged kernel can never hang the agent. Losing state on timeout is deliberate: interrupting one cell in-place would leave the interpreter in an unknown state.
- The kernel env is frozen at spawn (skills registering env passthrough later need `reset=true`); the schema description and config comment both say so.

**Wire protocol:** NDJSON requests on the kernel's stdin; responses framed on stdout behind a per-kernel random sentinel. Unframed bytes (fd-level output from subprocesses the user's code spawns) are attributed to the current cell — calls are serialized per kernel, so attribution is unambiguous. The generated RPC client reconnects once when `HERMES_RPC_PERSISTENT=1`, because a kernel legitimately outlives the RPC server's 300s idle window between cells; the host re-accepts for the kernel's lifetime.

## Related Issue

No existing issue proposes persistent execute_code state (searched open+closed; #77367 covers other harness patterns — vibe-mode workers — but not kernel persistence). Happy to open a tracking issue first if preferred.

## Type of Change

- [x] ✨ New feature (non-breaking change that adds functionality)

## Changes Made

- `tools/code_kernel.py` — new: kernel registry, runner source, frame reader, per-cell execution, teardown/atexit
[...]

PR #97265, opened 2026-08-28
fix(code-execution): key remote kernels by tool set

### LINKED ISSUE #97263: Remote session kernels reuse stale hermes_tools stubs after capability changes
## Summary

Persistent remote kernels are keyed by `(owner, "remote", env_type, task_env_id)`, while the generated `hermes_tools.py` module is built from the call's `sandbox_tools` only when that kernel is first spawned.

If the enabled sandbox tool set changes for the same owner/environment, the existing kernel is reused with stale Python tool stubs. Newly enabled tools cannot be imported until the kernel is explicitly reset. Removed tools can remain visible as stale stubs, although the per-cell RPC allow-list correctly rejects their execution.

## Current behavior

1. Start a remote kernel with `sandbox_tools={"read_file"}`.
2. Call it again for the same owner/environment with `sandbox_tools={"web_search"}`.
3. `_kernel_key()` returns the same key, so `_spawn_remote_kernel()` is not called again and `hermes_tools.py` is not regenerated.

The local session-kernel implementation already includes `tuple(sorted(sandbox_tools))` in its cache key and therefore does not have this drift.

## Expected behavior

The remote kernel key should include the sorted sandbox tool set, matching the local kernel invariant. A changed capability set should create a fresh kernel with matching stubs and per-cell RPC authority.

## Affected code

- `tools/code_kernel_remote.py::_kernel_key`
- `tools/code_kernel_remote.py::execute_in_remote_kernel`

This is a functional capability-refresh bug, not an RPC authorization bypass: the current per-cell allow-list remains authoritative.

Liuzikaii:
## Summary

- include the sorted sandbox tool set in persistent remote-kernel cache keys
- spawn a fresh kernel and regenerate `hermes_tools.py` when capabilities change
- preserve owner-based disposal while matching the local session-kernel key invariant
- add regression coverage for capability changes on the same owner/environment

Fixes #97263

## Testing

- `python -m pytest -p no:cacheprovider tests/tools/test_code_kernel_remote.py tests/tools/test_code_kernel.py -q`
- `python -m ruff check tools/code_kernel_remote.py tests/tools/test_code_kernel_remote.py`

### DIFF
--- a/tests/tools/test_code_kernel_remote.py
+++ b/tests/tools/test_code_kernel_remote.py
@@ -74,10 +74,11 @@ def _cell(status="ok", stdout="", execution_count=1, **kw):
     return payload
 
 
-def _run(env, code="print(1)", *, task="t1", reset=False, timeout=10):
+def _run(env, code="print(1)", *, task="t1", reset=False, timeout=10,
+         tools=frozenset({"read_file"})):
     return execute_in_remote_kernel(
         code, env=env, env_type="ssh", task_env_id=task,
-        sandbox_tools=frozenset({"read_file"}), timeout=timeout,
+        sandbox_tools=tools, timeout=timeout,
         max_tool_calls=5, reset=reset,
     )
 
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Keys remote kernels by tool set to avoid stale stubs.

**What it does**
- `tools/code_kernel_remote.py:_kernel_key` now `return (owner,"remote",env_type,task_env_id,tuple(sorted(sandbox_tools)))`; `execute_in_remote_kernel` passes `sandbox_tools` through. Switching from `{"read_file"}` to `{"web_search"}` spawns second kernel (`len(_REMOTE_KERNELS)==2`, two `nohup`), each with fresh tool stubs instead of reusing prior kernel's `read_file` stub when `web_search` needed.

**Non-blocking notes**
- `tuple(sorted(...))` makes key deterministic irrespective of input order — good.
- `_REMOTE_KERNELS` growth now bounded by tool-set permutations; still capped by existing `idle_exit` reaper and `max_session_kernels` eviction.

Correct isolation fix, same invariant as local (#94647).

kshitijk4poor:
Landed on `main` via #103668 as `7b8cf4c631`, with your authorship preserved on the fix commit, @Liuzikaii — thank you. Closing in favour of the merged salvage.

PR #97269, opened 2026-08-28
fix: fence lifecycle callbacks by generation and identity

ColDSnit:
## Summary
Fence delegated-work lifecycle callbacks by generation and callback identity so a stale or conflicting completion cannot replace the first accepted terminal result. Typed interruption evidence is first-writer-wins, with later observations kept only as bounded evidence.

## Test plan
- [x] `scripts/run_tests.sh tests/agent/test_subagent_lifecycle.py -q` — 8 passed, 0 failed

This change is in-memory only. It does not add durable restart reconciliation.

## Scope
Python lifecycle API and focused tests only. No desktop, diagnostics, or process-containment changes.

### DIFF
--- a/agent/subagent_lifecycle.py
+++ b/agent/subagent_lifecycle.py
@@ -52,7 +52,7 @@ class SubagentInterruptionCause(str, enum.Enum):
     """Authoritative reason associated with an interruption observation."""
 
     USER_CANCEL = "user_cancel"
-    USER = "user_cancel"
+    USER = "user_cancel"  # intentional alias of USER_CANCEL; same wire value
     TIMEOUT = "timeout"
     RESET = "reset"
     LEASE_LOSS = "lease_loss"
@@ -66,7 +66,7 @@ class SubagentInterruptionStage(str, enum.Enum):
 
     REQUESTED = "requested"
     SIGNAL_ACCEPTED = "signal_accepted"
-    SIGNAL = "signal_accepted"
+    SIGNAL = "signal_accepted"  # intentional alias of SIGNAL_ACCEPTED; same wire value
     OBSERVED = "observed"
     ACKNOWLEDGED = "acknowledged"
     VERIFIED_TERMINATION = "verified_termination"
@@ -705,6 +705,9 @@ class SubagentLifecycleService:
             record = _REGISTRY.records.get(handle.subagent_id)
             if record is None:
                 return None
+            # generation == 0 is a pre-field wildcard for serialized handles.
+            # It is not a freshness proof; authenticity is the capability HMAC
+            # and the active-parent session check above.
             if handle.generation not in (0, record.generation):
                 return None
             return record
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Sound hardening of delegated-work lifecycle; generation + callback identity fencing prevents stale/conflicting completions from clobbering first terminal result. First-writer-wins interruption evidence with bounded later-evidence is well scoped (in-memory only, no durable reconciliation claimed).

**What it does**
- `agent/subagent_lifecycle.py:48` adds `SubagentInterruptionCause` / `SubagentInterruptionStage` enums, `SubagentInterruptionEvidence` immutable bounded evidence, and extends `SubagentHandle`, `SubagentStatus`, `SubagentTerminalState`, `SubagentCancelResult`, `SubagentResult`, `_Record` with `generation`, `interruption`, `later_interruptions`, `dropped_interruptions`, `callback_digests`.
- `_MAX_INTERRUPTION_EVIDENCE=8` bounds later evidence; `_finalize_record` is generation-aware and idempotent, `_record_interruption_locked` enforces authoritative first-writer.
- Tests `tests/agent/test_subagent_lifecycle.py:182` cover typed first/later evidence, generation-aware idempotency, bounded eviction, and conflicting duplicate callback rejection.

**Non-blocking notes**
- `SubagentInterruptionCause.USER = "user_cancel"` duplicates `USER_CANCEL` — Python Enum makes `USER` an alias to `USER_CANCEL` (not a distinct member). Likewise `SubagentInterruptionStage.SIGNAL = "signal_accepted"` aliases `SIGNAL_ACCEPTED`. If alias intentional, add comment; if distinct, use unique values.
- `_Record.generation` defaults to `1` while `SubagentHandle.generation` defaults to `0` — launching sets `1`; older persisted handles without the field will deserialize with `0` vs new `1`. Verify no mismatch in persisted `to_dict`/`from_dict` rehydration.
- `first_cause`/`later_causes` repeated across 4 dataclasses — consider shared mixin/base to avoid drift.
- In-memory only is explicit and correct; no durable restart fence — good scope note.

ColDSnit:
Thanks for the review. Addressed the non-blocking notes in f2f9d66841 (rebased onto current `main`):

- **Enum aliases:** `USER` / `SIGNAL` are intentional aliases of `USER_CANCEL` / `SIGNAL_ACCEPTED` (same wire values). Added comments on the members so that is explicit.
- **Handle `generation` 0 vs record `generation` 1:** launching still writes generation 1. A handle serialized without the field deserializes to 0, which `_record()` treats as a pre-field wildcard. Authenticity remains the capability HMAC and active-parent check; 0 is not a freshness proof. Documented that on the guard. Added tests for `to_dict`/`from_dict` round-trip, missing-field → 0 accepted via public `status()`, and generation+1 rejected as `UNKNOWN`.
- **Shared `first_cause`/`later_causes`:** left as-is. The properties are trivial; a mixin would need two return shapes and would not fix a drift bug.
- **In-memory only:** unchanged; the existing service docstring already states there is no durable restart fence.

Local: `scripts/run_tests.sh tests/agent/test_subagent_lifecycle.py` → 11 passed.

PR #97270, opened 2026-08-28
fix: make approval and termination outcomes explicit

ColDSnit:
## Summary
Make approval and process-termination outcomes explicit and fail-closed for non-interactive work that has no approval surface. Interactive CLI `execute_code` keeps its historical per-call terminal guards. Dead PIDs are reported as already exited; only a live start-time mismatch refuses termination.

## Test plan
- [x] `scripts/run_tests.sh tests/tools/test_approval_process_safety_slice.py tests/tools/test_approval_outcome_parity.py -q` — 12 passed, 0 failed

## Scope
`tools/approval.py`, `tools/process_registry.py`, and focused tests. This does not add Windows Job Object or POSIX process-group launch containment.

### DIFF
--- a/tests/tools/test_approval_process_safety_slice.py
+++ b/tests/tools/test_approval_process_safety_slice.py
@@ -125,6 +125,7 @@ def test_identity_verified_tree_reports_survivors(monkeypatch):
             return [Child()]
 
     monkeypatch.setattr(ProcessRegistry, "_host_pid_is_ours", classmethod(lambda cls, pid, start: True))
+    monkeypatch.setattr(ProcessRegistry, "_is_host_pid_alive", classmethod(lambda cls, pid: True))
     monkeypatch.setattr(
         ProcessRegistry,
         "_safe_host_start_time",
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Focused security hardening slice — makes approval/termination outcomes machine-readable and fail-closed for non-interactive contexts, while preserving interactive CLI history. Well-tested and narrowly scoped.

**What it does**
- `tools/approval.py:193` adds `ApprovalOutcome` (str Enum) and `ApprovalBoundary` (frozen dataclass with `intent_digest`/`policy_digest` via SHA256), plus `set/reset_current_approval_operation` ContextVar and `_typed_approval_result`/`_typed_approval_guard` decorators to emit `outcome`, `retryable`, `operation_id`, `intent_digest`, `policy_digest` on legacy dict results.
- `tools/process_registry.py:352` adds `TerminationEvidence` dataclass and rewraps `_terminate_host_pid` to be identity-checked: `already_exited` for dead PID, `legacy_unverified` when no baseline, `identity_unavailable`/`identity_mismatch` on start-time mismatch (never calls `_terminate_host_pid_legacy`), snapshots descendant PIDs via `psutil` before kill and reports `partial` vs `terminated` with `targeted_pids`/`survivors`.
- Tests `tests/tools/test_approval_process_safety_slice.py:37` cover background guard non-bypass, frozen boundary, `POLICY_DENIED`, `TIMEOUT` vs `DENIED` distinct, identity-mismatch never reaches terminator, verified tree survivors, interactive-CLI historical path.

**Non-blocking notes**
- `_typed_approval_result` infers `NON_INTERACTIVE` vs `DENIED` via substring checks on `message` (`"no interactive"/"non-interactive"` → `NON_INTERACTIVE` else `DENIED`). Fragile if message phrasing changes; prefer typed source where possible.
- `policy_digest` uses `json.dumps(..., default=str)` — stable for current dict but `default=str` can be non-deterministic for custom objects; fine for present caller but document.
- `ApprovalBoundary.for_request` hashes `{"command": command, "env_type": env_type}` — command may contain secrets; hashing avoids persistence but ensure no logging of raw command elsewhere.
- Process snapshot is best-effort race: children spawned between snapshot and `_terminate_host_pid_legacy` won't be in `targets`; post-kill survivor check partially mitigates — acceptable given no new supervisor.
- `retryable = bool(approved)` makes only approved retryable; timeout/denied correctly non-retryable without new `operation_id` — matches stated intent.

ColDSnit:
Thanks for the review. Addressed the non-blocking notes in 796525b8c6 (rebased onto current `main`):

- **Substring `NON_INTERACTIVE` vs `DENIED`:** producers that already set `outcome` still win. The message-substring path is now documented as last-resort for untyped legacy dicts only. Added tests that a typed `NON_INTERACTIVE` is preserved even when the message would look like a generic deny, and that an untyped generic deny stays `DENIED`. Public background-guard coverage already asserts `NON_INTERACTIVE` on `check_all_command_guards` / `check_execute_code_guard`.
- **`policy_digest` `default=str`:** removed. Policy for this slice is JSON-safe `{mode, cron, single_query}`; a non-JSON-safe object now fails at dumps time instead of hashing `str(obj)`. Digest stability is tested for key-reordered primitive dicts.
- **Command hashing vs logging:** `for_request` still hashes command+env and does not persist the command. No raw-command logger in this slice; a repo-wide log audit is out of scope here.
- **Snapshot vs spawn race:** left as best-effort (as you noted). Comment on the snapshot block now says children spawned after the snapshot are outside `targets`.
- **`retryable`:** unchanged; still `bool(approved)`.

Local: `scripts/run_tests.sh tests/tools/test_approval_process_safety_slice.py` → 13 passed.

PR #97271, opened 2026-08-28
feat: add a bounded local diagnostic bundle

ColDSnit:
## Summary
Add `hermes debug bundle`, a local-only, allowlisted, always-redacted diagnostic collector. The bundle path does not upload, does not accept `--no-redact`, and writes atomically under the active profile diagnostics directory. Log snapshots keep complete lines when truncated on LF or CRLF.

## Test plan
- [x] `scripts/run_tests.sh tests/hermes_cli/test_debug_bundle.py tests/hermes_cli/test_debug.py -q` — 56 passed, 0 failed

Existing `hermes debug share` behavior is unchanged aside from dispatching the new `bundle` subcommand.

### DIFF
--- a/hermes_cli/debug.py
+++ b/hermes_cli/debug.py
@@ -802,6 +802,9 @@ def _diagnostic_truncate_text(value: str, max_bytes: int) -> tuple[str, bool]:
     encoded = value.encode("utf-8")
     if len(encoded) <= max_bytes:
         return value, False
+    # Loss-tolerant UTF-8 prefix: drop at most one partial trailing codepoint.
+    # errors="ignore" is deliberate so a U+FFFD replacement cannot enlarge the
+    # already-truncated value against the global diagnostic byte budget.
     return encoded[:max_bytes].decode("utf-8", errors="ignore"), True
 
 
--- a/hermes_cli/debug.py
+++ b/hermes_cli/debug.py
@@ -15,6 +15,8 @@ Currently supports:
                           (viewable only by Nous staff / allowlisted mods via
                           a Google-login-gated viewer) and auto-deletes after
                           14 days, rather than going to a public paste.
+    hermes debug bundle   Write a bounded, redacted diagnostic bundle locally;
+                          this path never uploads or accepts ``--no-redact``.
 """
 
 import datetime
@@ -22,8 +24,11 @@ import gzip
 import io
 import json
 import logging
+import os
+import platform
 import re
 import sys
+import tempfile
 import time
 import urllib.request
 from dataclasses import dataclass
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Well-scoped local diagnostic surface that stays bounded, always-redacted, and profile-scoped. Separation from the legacy upload path (`collect_share_bundle` / `--no-redact`) is clean and correctly never inherits the escape hatch.

**What it does**
- `hermes_cli/debug.py:88` adds `hermes debug bundle` (no `--no-redact` flag) and ~500 lines of local bundle logic: schema `hermes-debug-bundle/1`, allowlisted logs (`agent/errors/gateway/gui/desktop`), limits (`96KB` input, `24KB` section, `160KB` total, `200` lines), redaction via `_redact_log_text` + collapsing `«redacted:…»` to `[REDACTED]`, UTF-8-safe truncation, atomic `os.replace` via temp file, and strict containment via `_diagnostic_safe_root`/`_diagnostic_path_is_safe` (symlink/reparse rejection, `resolve(strict=True)` on root, `relative_to` checks).
- Fixes log snapshot truncation `hermes_cli/debug.py:71` to keep complete lines: `rfind(b"\n",0,cut)+1` retains remainder of first line (handles CRLF midpoint) rather than dropping it; exceeds budget by at most one line — sound tradeoff.
- `hermes_cli/subcommands/debug.py:171` wires parser; tests `tests/hermes_cli/test_debug_bundle.py:214` verify no unredact option, redaction before serialization, bounds enforcement, profile-scoped local-only behavior, and staging cleanup on `os.replace` failure.

**Non-blocking notes**
- `_DIAGNOSTIC_RESERVED_BASENAMES` covers Windows reserved names (`CON/PRN/AUX/NUL/COM/LPT`) — good, but only checked against basename stem; ensure extension handling (`CON.txt` still blocked) if path sanitizer uses full basename.
- `_diagnostic_truncate_text` decodes with `errors="ignore"` — avoids split multi-byte; consider ensuring truncated value still ends at valid char boundary (current slice may drop tail bytes, `ignore` handles it but `errors="replace"` would be more explicit).
- `urllib.request` imported but unused in bundle path in this slice — verify no leftover upload code path reachable when `bundle` command errors.
- Atomic commit cleans `*.tmp` on failure — good; ensure temp file is created with `mkstemp` under diagnostics dir (not system tmp) to keep `replace` atomic on same filesystem.

ColDSnit:
Thanks for the review. Addressed the non-blocking notes in 15e5fe768d (rebased onto current `main`):

- **Windows reserved names / `CON.txt`:** `_diagnostic_valid_output_name` already inspects the stem before the first dot (`CON.txt` → `CON`). Added a regression through `_diagnostic_output_path` for `CON.txt`, `con.txt`, and `AUX.log`.
- **Truncation `errors="ignore"`:** kept. The helper returns a loss-tolerant UTF-8 prefix (at most one partial trailing codepoint dropped) so a U+FFFD replacement cannot grow the global byte budget. Comment + bounded UTF-8 test added.
- **`urllib.request`:** still used by the share/upload helpers in the same module. The bundle path does not call it; existing `test_bundle_is_local_only_and_profile_scoped` already asserts `urlopen` is not used.
- **Atomic `mkstemp`:** already uses `dir=destination.parent`. Added a spy test that the staging directory is the diagnostics dir, not `tempfile.gettempdir()`.

Local: `scripts/run_tests.sh tests/hermes_cli/test_debug_bundle.py` → 8 passed.

PR #96726, opened 2026-08-27
Rebuild Bot Mode on the app's design system

OutThisLife:
## Summary

Bot Mode arrived as a 16,291-line `plugin.js` that reimplemented most of the desktop app instead of using it: its own scroll container, color palette, status dots, empty states, buttons, time formatting and cron surface, none of which could follow the theme. It also shipped a second "Bots" workspace tab that duplicated the sidebar, and a workspace-scoping filter that could take the main zone off screen entirely.

This rebuilds it on the primitives the app already ships, and widens the plugin SDK with the ones it was forced to reinvent so the next plugin doesn't have to.

- **Bot Mode is now 44 focused modules** (`plugin.tsx` is imports and `register()`). A bot row renders the same `SessionStatusDot`, swatches and age labels as the session row beside it; the UI settles on "bot" as the noun, with model-facing prompt text still saying "agent".
- **It speaks every language the app does.** The ~170 Bot-Mode-specific strings live in a plugin locale bundle registered through `ctx.i18n`, shipped in `en` / `ja` / `zh` / `zh-hant` using core's own nouns (ゲートウェイ / 网关 / 閘道, プロファイル / 配置文件 / 設定檔) so the rail doesn't read like a second translation of the app. Vocabulary core already ships in every locale — generic verbs, weekday names, Daily/Hourly — resolves against core rather than being duplicated into a second, worse translation. That includes the intro a new bot is born with, which used to greet every user in English and pull the bot's reply into English with it.
- **The SDK exports what was being hand-rolled** — the `Panel` master-detail family, `SessionStatusDot`, `ColorSwatches` + `PROFILE_SWATCHES`, `RowButton`, `DisclosureCaret`, `formatAgo`, `translateNow`, and the unread store behind the dot — each with a docblock naming the hand-rolled shape it replaces.
- **The main zone can no longer vanish.** Workspace mode goes back to being a signal consumers adapt to, never a decision about whether a pane renders, and a structural floor keeps the registered main pane's subtree alive whatever happens to its tabs. Bot chats are ordinary tabs in the main strip. The duplicate "Bots" workspace tab and its auto-open/close signal machine are gone (-1,623 lines).
- **Bot chats share core's unread store** rather than keeping a parallel map that could disagree with the session dot. Hidden sessions have no listed row to read a profile from, so the writers take an explicit profile hint.
[...]

Issue #95044, opened 2026-08-25
Desktop: tab switches stay slow on huge sessions — residual cost after #55191 (full-store hydration, unpaginated sidebar, uncollapsed tool rows)


tarttee:
## Environment
- Hermes Desktop build 2026-08-23 (desktop-build-stamp), Windows 11, Electron renderer 1,114 MB private / 3,179 CPU-sec after 42 h uptime.
- state.db: 2,811 sessions, 141,134 messages, 284 MB of stored tool results (read_file alone 22,631 calls / 122 MB).

## Symptom

Tab/session switches from the left sidebar hitch on sessions with thousands of tool rows. The heaviest session has 2,642 tool rows / 6.44 MB; a single autonomous-run day (cron loops + delegations) produced 2,379 sessions / 57,021 tool rows / 180 MB.

## What already exists (good)

- #55191 transcript windowing (`transcript-window.ts`) bounds what assistant-ui materializes by render weight, with a sticky cut and the DOM budget in `thread/list.tsx`.
- Sidebar list is virtualized (`virtual-session-list.tsx`).
- Sidebar API defaults `archived=exclude`; backend `sessions.auto_archive` sweep exists (opt-in).

## What is still O(n) on switch

1. **Full-session hydration into the store.** Opening a heavy session still loads and normalizes every message into the runtime repository before windowing (`thread-loading.ts`, `session-state-cache.ts`). Windowing caps materialization, not hydration. For a 2,600-tool-row session that is a big parse on every open.
2. **Sidebar endpoint returns every session.** The virtualized list still fetches all 2,811 rows from `/api/profiles/sessions/sidebar`; cursor/pagination would cut per-switch payload and re-render churn.
3. **Tool rows render full content in the visible window.** In-window historical tool rows carry multi-KB payloads through markdown/JSON normalize + store weight. A collapsed-by-default tool row (summary line + expand) would both cut render weight and make `TRANSCRIPT_WINDOW_BUDGET` go 4-5x further.

## Suggested shape (respecting the narrow waist)

- Hydrate per-session lazily: fetch the tail window first, fetch earlier on "Show earlier" (window + backfill already exist client-side; thread them into loading).
- Paginate the sidebar endpoint (cursor), keep the virtualized DOM.
- Render tool rows collapsed by default in historical messages, expanding on click; weight function discounts collapsed rows.
- Optional: a bulk "archive heavy sessions" affordance (I shipped a local plugin for this; core affordance would be nicer).

[...]

PR #89466, opened 2026-08-18
feat(desktop): Sessions list shows every connected gateway's chats (#88880)

teknium1:
## Summary

The global SESSIONS sidebar now aggregates every connected gateway's chats — local profiles, v1 per-profile remote overrides, AND v2 registry connections (remote/cloud/ssh). Previously a session created on a registry-connected remote Bot persisted fine on the remote backend but never appeared in the Sessions list, and the renderer received an empty array despite a successful API response (#88880). Bot Mode's hidden canonical chats stay excluded from the global list, exactly as before.

Root cause: the Electron unified-list splice (`mergeRemoteProfileSessions`) only knew v1 per-profile remote overrides. Rows on v2 registry gateways were never fetched, so the aggregate simply didn't contain them.

## Changes

- `electron/profile-session-routing.ts`: `fetchRegistrySessionRows` reads each CONNECTED registry gateway's sessions (ssh backends natively; shared remote/cloud hosts via one cross-profile aggregate, with a flat-list fallback for older backends), tagging rows with `connection_id` + owning `profile`. `spliceRegistrySessionRows` dedupes into the unified list and extends per-profile totals. Reads never pass `include_hidden` — hidden Bot Mode chats stay hidden.
- `electron/main.ts`: unified-list + batched-sidebar intercepts fire when registry gateways are pooled; only already-pooled backends are read (a sidebar refresh never dials/spawns — the roster-respawn trap), and a dead gateway contributes nothing.
- `types/hermes.ts` + `use-session-actions`: `SessionInfo.connection_id`; resuming a registry-owned row activates its connection-scoped gateway (`ensureGatewayAgent`) instead of a same-named local profile.
- `store/gateway.ts`: `ensureActiveGatewayOpen` rides out an in-flight secondary activation (bounded 8s) instead of failing instantly — the Sessions `+` during remote wake no longer errors "Hermes gateway is not connected" (#88880's secondary symptom).

## Validation

| Check | Result |
|---|---|
| New registry-source/splice unit tests (tagging, shared-host + legacy fallback, dead-gateway isolation, hidden-flag contract, dedupe/totals) | 5/5 pass |
| Activation-wait regression test (sabotage run: fails on pre-fix gateway.ts) | verified both ways |
| electron/ + src/store/ + hermes.ts + session hooks vitest | 2,981 passed |
| `npm run check:lint` (3 tsconfigs + eslint) | 0 errors |

Closes #88880.

## Infographic

All gateways, one sessions list

Issue #88880, opened 2026-08-18
Desktop v2 remote Bot sessions stay absent from sidebar despite successful remote API response


kiyo-e:
## Summary

In Hermes Desktop with both a local connection and a v2 remote registry connection, selecting a remote Bot and creating a chat can persist a valid remote `desktop` session while the global **SESSIONS** section continues to render **No sessions yet**.

This is not a missing-data or hidden-session problem: the remote sidebar API returns the rows, including the newly created Desktop sessions, but the renderer's Sessions component still receives an empty `sessions` array.

The Sessions `+` action can also fail with `Hermes gateway is not connected` if it is used while the remote profile socket is still activating/reconnecting.

## Environment

- Hermes Agent/backend: `v0.20.2`
- Desktop source base: `main` at `4323c67dcc6048fc8e311cdff7600d3d6a17807f`
- macOS arm64
- One local connection plus one v2 remote connection
- Remote source exposes multiple profiles/Bots

## Reproduction

1. Configure a local Desktop connection and a v2 remote connection.
2. Open Bot Mode and select a Bot owned by the remote connection.
3. Choose **New chat with this agent**.
4. Send a message and wait for the turn to persist.
5. Inspect the global **SESSIONS** section in the left sidebar.
6. Optionally click the **SESSIONS** `+` action while the remote profile is still waking/reconnecting.

## Actual behavior

- The chat is persisted on the remote backend.
- The global Sessions section remains `No sessions yet`.
- The Sessions `+` action may report `Hermes gateway is not connected` during remote activation.
- Bot-specific/canonical navigation and the global session list can therefore disagree about whether the remote chat exists.

## Expected behavior

After the first message is persisted, the newly created remote Desktop session should appear in the global Sessions section. Session creation should either wait for the selected remote gateway to become ready or remain disabled while that gateway is not connected; it must not silently target a same-named local profile.

## Direct evidence

The remote backend contains recent rows such as:

- `profile: hermes-claude`
- `source: desktop`
- `message_count: 2`
- `hidden: 0`

The exact batched sidebar request used by Desktop was observed going through the intended non-primary registry connection:

```text
/api/profiles/sessions/sidebar?recents_profile=all&recents_limit=50&...
```

Calling that request through the same Desktop bridge returned:

```text
recents.sessions.length = 50
errors = []
```

The returned recents included unhidden `desktop` sessions for the selected remote profile. At the same time, the rendered Sessions component had:

```text
sessions.length = 0
[...]

PR #97273, opened 2026-08-28
fix(desktop): bound cross-gateway session list reads

dokterdok:
## Keep Long Session Lists Visible

The unified Sessions view can silently lose a gateway's rows when Desktop requests more than 100 sessions. Load-more and owner lookups can exceed that window, but the flat `/api/sessions` endpoint rejects limits above 100 with `422`; the failed source then contributes no rows.

This PR reads API-safe pages and reassembles the requested window before attaching its connection and profile. The same helper covers registry sources, the legacy shared-host fallback and direct remote-profile reads. No backend or schema change is needed.

## Failure Behavior

A later-page failure rejects that backend's requested window rather than returning an unmarked partial list. Registry callers retain their existing per-backend fallback, so healthy gateways still contribute. A subsequent attempt starts from the original offset and refetches the complete window.

This does not promise that every cached row survives an outage. Explicit per-source completeness and cache retention would require a separate change; silently returning a successful prefix is not that change.

## Scope

- No change to the existing `profile=all` aggregate-profile contract, hidden sessions or Bot Chat ownership.
- No new endpoint, persisted state or automatic partial-result API.
- No full transcript hydration or sidebar virtualization changes; #95044 tracks that broader performance class.

## Related Work

- Merged #89466 introduced cross-gateway registry reads for #88880; this PR bounds their flat fallback to the existing backend limit.
- #81457 owns adjacent authority/profile selection. Both touch `profile-session-routing.ts`; whichever lands second must preserve both behaviors.

## Validation

Current `54d05bc4101` adds a contract comment and three failure/recovery cases to the existing suite. **24 routing tests pass**, covering direct remote, registry SSH and legacy fallback, with healthy-source retention and fresh recovery after second-page failure. Emitted production JavaScript is unchanged from `eb7c447dbc`. Targeted lint, formatting and diff checks pass; changed source/test owners remain below 2,000 lines. Current CI is tracked separately.

<details>
<summary>Earlier acceptance at eb7c447dbc</summary>

Recorded on post-#96726 main `7eee066c30`:

- Focused routing and REST-helper selection: 57 tests across two files passed.
- Electron selection: 1,911 passed, 6 skipped.
- Desktop renderer, Electron and E2E typechecks passed.
- Targeted lint passed; full Desktop lint had zero errors and 124 existing warnings.
- A broader Electron run reached 1,938 passed and 6 skipped, while reproducing the pre-existing macOS `/bin/true` fixture failure on exact main. That portability correction remains outside this PR.

These historical selections overlap, are not added together, and are not presented as a fresh full-suite run of the new test-only follow-up.

</details>

### DIFF
--- a/apps/desktop/electron/profile-session-routing.test.ts
+++ b/apps/desktop/electron/profile-session-routing.test.ts
@@ -302,6 +302,38 @@ test('registry sources: shared remote hosts read the cross-profile aggregate onc
   )
 })
 
+test('registry sources page flat session endpoints at their 100-row ceiling', async () => {
+  const calls: string[] = []
+  const sessions = Array.from({ length: 150 }, (_, index) => ({ id: `session-${index}` }))
+
+  const rows = await fetchRegistrySessionRows(
+    [
+      {
+        connectionId: 'gw-ssh',
+        kind: 'ssh',
+        backends: [{ descriptor: 'ssh-desc', profileLabel: 'default' }]
+      }
+    ],
+    new URLSearchParams({ limit: '150', offset: '0', profile: 'all' }),
+    async (_descriptor, path) => {
+      calls.push(path)
+      const url = new URL(path, 'http://desktop.test')
+      const limit = Number(url.searchParams.get('limit'))
+      const offset = Number(url.searchParams.get('offset'))
+
+      assert.ok(limit <= 100)
+
+      return { sessions: sessions.slice(offset, offset + limit), total: sessions.length, limit, offset }
+    }
+  )
+
+  assert.deepEqual(calls, [
+    '/api/sessions?limit=100&offset=0',
+    '/api/sessions?limit=50&offset=100'
+  ])
+  assert.equal(rows.length, 150)
+})
+
 test('registry-pinned session responses retain their owning connection', () => {
   const sidebar = tagRegistrySessionResponse(
     '/api/profiles/sessions/sidebar?recents_profile=default',
--- a/apps/desktop/electron/profile-session-routing.test.ts
+++ b/apps/desktop/electron/profile-session-routing.test.ts
@@ -224,6 +224,112 @@ test('remote session reads keep small requests on one call', async () => {
   assert.equal(result, expected)
 })
 
+test('a later remote page failure rejects the window and recovery fetches it afresh', async () => {
+  const params = new URLSearchParams({ profile: 'remote-work', limit: '150', offset: '0' })
+  const calls: string[] = []
+  const failure = new Error('second page unavailable')
+  let failing = true
+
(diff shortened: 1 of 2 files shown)


Enough1122:

Good refactor. Extracting the paging loop into `fetchPagedSessionList` and routing all three `/api/sessions` call sites through it removes duplication and fixes a real 406/truncation hazard (a single `limit=150` request against a 100-row ceiling). The test that asserts the `limit <= 100` page ceiling and the exact call sequence (`limit=100&offset=0` → `limit=50&offset=100`) is exactly the right contract to pin. I also checked the no-limit case: when the caller passes no `limit`, `Number(null)` is `NaN` so `needsPaging` is false and behavior is unchanged — good backward-compat.

A few points to consider:

1. **A later-page failure now discards all rows (partial-loss regression risk).** In the registry path the helper call is wrapped in `.catch(() => null)` (profile-session-routing.ts, both new call sites). If page 1 succeeds but page 2 throws (flaky SSH/remote backend), the whole result is dropped and the caller renders "no sessions" even though it already had 100 valid rows. The single-call code never had this shape — it either got everything or null. Consider catching inside the paging loop and returning the accumulated `firstPage`/partial rows as best-effort when a non-first page fails, instead of letting the exception escape to the outer `.catch`. (A first-page failure can still return null.)

2. **Paging multiplies round-trips per remote backend.** The shared-remote aggregator loop now issues `ceil(limit/100)` sequential requests per source. For a sidebar "recent sessions" request the default limit is presumably modest, so this is usually one call — but if any caller passes a large `limit` against several SSH backends, you go from N calls to N×ceil(limit/100) serial round-trips. Worth confirming the desktop callers' default limits are small, and/or documenting that the helper is deliberately serial to keep per-backend ordering and the `total` from `firstPage`.

3. **`needsPaging` treats any `limit > 0` as pagable** — including tiny values (e.g. `limit=5`), which still runs the paging branch (harmless, one 5-row page). If you want the ceiling math to be the only special path, gate `needsPaging` on `requestedLimit > 100`; either way the behavior is correct, just slightly clearer intent.

Minor: the two new registry call sites in `fetchRegistrySessionRows` are near-identical wrappers — if a third site appears, a tiny helper that takes `(descriptor, params)` and does the `.catch(() => null)` + `tag()` would collapse them. Not needed yet, though.

Overall this is a clean, well-tested change.

dokterdok:
Thanks. I kept complete-or-fail paging per backend: returning an unmarked partial window could misrepresent the result. Healthy gateways still contribute through the existing fallback. 54d05bc4101 adds failure/recovery coverage for direct, SSH-registry and legacy paths; all 24 routing tests pass. Keeping every cached row during an outage needs a separate source-status/cache policy and is not claimed here. Small requests already stay on one call; Number(null) is 0.

PR #97276, opened 2026-08-28
fix(auth): allow profiles to share credential stores

J-Patch:
## Summary
- add strict `auth.source_profile` routing so an execution profile can use one exclusive local credential store
- pin one immutable auth authority across complete sync/async login, logout, status, runtime refresh, cache, pool mutation, cleanup, suppression, and dashboard operations
- reject malformed/self-referential/traversal/symlink/non-regular source paths and fail closed without root fallback
- bind loaded credential-pool objects to their original authority for later mutations and OAuth refresh/quarantine

## Verification
- 114 focused auth/OAuth/credential-pool tests passed on the exact candidate
- independent repository-wide review: GO, 192 tests passed, no medium+ correctness findings
- compileall and `git diff --check` passed
- hostile B→C fixtures cover reads, locks, writes, refresh, status, normalization, cleanup, suppression, CLI, and async dashboard handlers

No credential values are copied between profiles and no default/global profile is changed.

### DIFF
--- a/agent/credential_pool.py
+++ b/agent/credential_pool.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from functools import wraps
 import logging
 import os
 import random
@@ -43,6 +44,15 @@ from hermes_cli.auth import (
 )
 
 logger = logging.getLogger(__name__)
+
+
+def _pin_pool_auth_authority(operation):
+    """Run a pool method against the authority captured when it was loaded."""
+    @wraps(operation)
+    def pinned(self, *args, **kwargs):
+        with auth_mod._use_auth_authority(self._auth_authority_binding):
+            return operation(self, *args, **kwargs)
+    return pinned
 
 
 def _load_config_safe() -> Optional[dict]:
@@ -712,6 +722,7 @@ def _write_through_provider_state_to_global_root(
 
 class CredentialPool:
     def __init__(self, provider: str, entries: List[PooledCredential]):
+        self._auth_authority_binding = auth_mod._capture_auth_authority()
         self.provider = provider
         self._entries = sorted(entries, key=lambda entry: entry.priority)
         self._current_id: Optional[str] = None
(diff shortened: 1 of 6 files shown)


Enough1122:

**Overall:** High-sensitivity auth routing change that correctly keeps `auth.source_profile` as one exclusive, strictly-validated credential store with fail-closed semantics and authority pinning across sync/async boundaries. Pool objects stay bound to the authority at load time to prevent OAuth refresh split-brain.

**What it does**
- `hermes_cli/auth.py:190` adds `_configured_auth_source_path()` (validates `auth.source_profile` as simple `[-_a-zA-Z0-9]` name, resolves via `profiles_root.resolve(strict=True)`, rejects traversal/symlink/non-regular/self-reference/escapes, requires directory existence, rejects symlink `auth.json`), plus `_AUTH_AUTHORITY_LOCAL` (threading.local), `_resolve_auth_authority()` and `_auth_authority_context()` to pin one resolved path for a complete operation, and `_pin_auth_authority` decorator (handles `iscoroutinefunction`).
- Applies `@_pin_auth_authority` to `auth_add/remove/reset`, `_resolve_zai_base_url`, `load_pool`, web `add/remove pool` handlers; `agent/credential_pool.py:45` stores `_auth_authority_binding` at `CredentialPool.__init__` and re-enters `_use_auth_authority` in `_persist`/`_sync_*`/`_refresh_entry`/`_replace_entry` path (pool stays on original authority even if config changes mid-life).
- `write_credential_pool` writes and lock paths now follow source profile (`_auth_file_path`/`_auth_lock_path`/`_global_auth_file_path==None` when configured); `read_credential_pool` and provider state reads follow same authority.
- Tests `tests/hermes_cli/test_auth_profile_fallback.py:367` cover shared read/write + lock colocation, traversal/symlink/self-reference rejection, symlink file rejection, real-home guard during pytest, and fail-closed on malformed/unreadable config.

**Non-blocking notes**
- Large mechanical surface (6 files, 741 lines test) — well-covered by 114 focused + 192 repo-wide tests noted; keep future changes to `_pin_auth_authority` minimal as decorator covers both sync and async.
- `replace("-", "").replace("_", "").isalnum()` allows e.g. `--__` or `_` only? Actually `-`/`_` stripped then `isalnum` requires at least one alphanum after stripping — `---` would be rejected (empty after replace → `False`). Good.
- `resolve(strict=True)` on `profiles_root` will throw if profiles dir not yet created — caller correctly surfaces as RuntimeError fail-closed; ensure first-run `profile create` creates root before `auth` reads.
- `threading.local` authority pinning is per-thread; async handlers run on event-loop thread — single authority per operation still holds, but concurrent async ops on same thread will share local until context exits — `_auth_authority_context` stacks `previous` correctly. Keep nesting tested (seen `previous_home` check).
- No credential values copied between profiles — explicit and correct; document operator expectation that `coder` profile must have `auth.source_profile: work` and not attempt manual union.

Issue #91789, opened 2026-08-21
[Bug] zai: glm-5.3 accepts graded reasoning_effort (low/medium/high/max) but provider plugin only sends binary thinking — _is_glm_5_2() misses glm-5.3


terje1965:
## Bug Description

`glm-5.3` via the zai provider accepts **graded** `reasoning_effort` values at the API level, but Hermes never sends them — the zai provider plugin only emits binary `thinking: enabled/disabled` for glm-5.3, so `agent.reasoning_effort: low` (or medium) is silently dropped.

## Evidence (measured 2026-08-21, endpoint `api.z.ai/api/coding/paas/v4`)

Same non-trivial prompt sent raw with each effort level, reading `reasoning.tokens_details` from the response:

| effort sent | reasoning tokens returned |
|---|---|
| (none / default) | 69 |
| `low` | 4 |
| `medium` | 11 |
| `high` | 98 |
| `max` | 125 |

- No HTTP 400 for any level — the endpoint accepts `low` through `max` on `glm-5.3`.
- Token scaling is monotonic, i.e. these are real graded controls, not ignored fields.

## Root Cause

1. `plugins/model-providers/zai/__init__.py:49` — `_is_glm_5_2(model)` only matches glm-5.2 spellings. For `glm-5.3` it returns `False`, so the branch that emits `reasoning_effort` is never taken; only the binary `thinking` object is sent.
2. `agent/reasoning_effort.py:117` — `GLM52_EFFORTS = ("high", "max")` reflects GLM-5.2's documented levels, but the measurement above shows glm-5.3 accepts at least `low` and `medium` as well.

## Steps to Reproduce

1. Configure zai provider with `glm-5.3` (Coding plan endpoint).
2. Set `agent.reasoning_effort: low` in config (or via /model or profile override).
3. Observe outgoing request payload: no `reasoning_effort` field is emitted for glm-5.3; only `thinking: {type: enabled}`.
4. Send the same request with `reasoning_effort: "low"` raw to the endpoint — accepted, with reduced reasoning tokens as shown above.

## Expected Behavior

Graded effort levels configured in Hermes should reach the API for glm-5.3: at minimum the existing clamp should recognize glm-5.3 as a member of the GLM-5.x family, and the allowed-levels tuple should be verified per-model (5.3 demonstrably accepts low/medium in addition to high/max).

## Actual Behavior

Any effort setting other than off is silently coerced to default thinking; no error, no effect.

## Environment

[...]

PR #96849, opened 2026-08-28
fix(zai): drop reasoning_effort medium on the glm-5.3 flash SKU

liuhao1024:
## What does this PR do?

Stops `glm-5.3-flash` turns from dying at the wire with HTTP 400 code 1210 when the session's reasoning config carries `medium`.

The base `glm-5.3` accepts the four-step graded effort scale (`low`/`medium`/`high`/`max`, verified live in #91789), but its **flash SKU** rejects `medium` — the error text names its actual ladder: "please use low, high, or max". The declared `GLM53_EFFORTS` vocabulary applied uniformly to every 5.3 model, so any session whose reasoning config carried `medium` (stale session snapshot or explicit config — both observed in #96838) failed on every turn.

The fix narrows the flash SKU's ladder to `low`/`high`/`max` inside the GLM-5.3 branch of the effort mapping and declares `medium → low` as an override (nearest-weaker, mirroring the China-endpoint precedent in this provider). The base model keeps the full graded scale. This also absorbs the reported symptom's payload: even a stale `medium` now lands as `low` instead of a 400.

Scope note: #96838 also describes a deeper defect — the request path building from `_session_init_model_config["reasoning_config"]` captured at session creation rather than the freshly switched config. That is comp/agent session-state territory and is not addressed here; this PR fixes the provider-side vocabulary so the declared effort is always servable.

## Related Issue

Fixes #96838

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `plugins/model-providers/zai/__init__.py` — in `_glm_5_2_reasoning_effort`'s GLM-5.3 branch, detect the flash SKU and narrow the ladder to `low`/`high`/`max` with a declared `medium → low` override.
- `tests/plugins/model_providers/test_zai_profile.py` — parametrized regression for the flash SKU ladder (medium→low, the rest unchanged) plus a guard that the base `glm-5.3` still accepts `medium`.

## How to Test

1. `python -m pytest tests/plugins/model_providers/test_zai_profile.py -q` — should pass (47 passed).
2. Red/green verified locally: with the pre-fix mapping restored, the flash `medium → low` case fails (emits `medium`, which the SKU rejects with 1210); with this change all pass.
3. Manual equivalent: `hermes --provider zai --model glm-5.3-flash` with `agent.reasoning_effort: medium` (or a session created under that config) — turns should succeed instead of HTTP 400 code 1210.

### Code

[...]

PR #97286, opened 2026-08-28
fix(zai): glm-5.3 rejects medium and the thinking off switch (1210)

ivfarias:
## What does this PR do?

Two request shapes the `zai` profile sends today are rejected outright by `https://api.z.ai/api/paas/v4` — the `base_url` this profile ships — on `glm-5.3-flash`:

```
1210: This model always engages in thinking and cannot be disabled; please use low, high, or max
```

**1. `reasoning_effort: medium` fails every turn.** #91789 declared `GLM53_EFFORTS = ("low", "medium", "high", "max")` from probes against `api.z.ai/api/**coding**/paas/v4`. That route really does accept `medium` (re-confirmed below), but the standard route does not — so since #91789 landed, anyone on `glm-5.3*` with the most common effort level gets a hard API error on every request.

**2. The thinking off switch fails every turn.** `parse_reasoning_effort("none")` (also `false`/`off`/`disabled`, and the desktop toggle) yields `{"enabled": False}`, and `build_api_kwargs_extras` turns that into `thinking: {"type": "disabled"}`. GLM-5.3 has no off switch and rejects the marker. Simply dropping it isn't right either: omitting the field leaves the server default, which is `max` — the *most* expensive tier — so "off" would cost more than any level the user could have asked for.

## Related Issue

No upstream issue. Observed on a production deployment running `glm-5.3-flash` through this profile, then isolated with the live probes below.

## Type of Change
- 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- **`agent/reasoning_effort.py`** — `GLM53_EFFORTS = ("low", "high", "max")`. Pure data change, per the module's own rule 3 ("when a provider rejects a level, fix its declared supported set"). `clamp_effort` then resolves `medium` to `low` (nearest weaker) with no call-site logic. The comment records both routes and why the strict one is the safe vocabulary to declare.
- **`plugins/model-providers/zai/__init__.py`**
  - `_glm_5_2_reasoning_effort`: an explicit disable resolves to `"low"` on GLM-5.3 — the cheapest level the wire accepts — instead of `None`. GLM-5.2 keeps returning `None` (it has a real off switch).
  - `build_api_kwargs_extras`: the `thinking` disable marker is not emitted for GLM-5.3. `thinking: {"type": "enabled"}` is still sent and still accepted.
  - Module and `_is_glm_5_3` docstrings updated.
- **`tests/plugins/model_providers/test_zai_profile.py`** — extends the existing `TestZaiGLM53ReasoningEffort`: the `medium` row of the parametrized mapping now expects `low`, plus one new test for the disable path.

GLM-5.2 and GLM-4.5 are untouched — 5.2 still clamps `low → high` and still sends a genuine disable marker (`TestZaiGLM52ReasoningEffort` unchanged and green).

### Trade-off

Users pointed at the coding route lose `medium` and get `low` instead — one notch of reasoning depth, on the side of the module's stated "never silently escalate cost above what was asked" rule. Users on the shipped default route stop getting a hard 1210 on every request. If per-route vocabularies are wanted later, that needs a `base_url`-aware selector; this PR does not add one.

## How to Test

1. `pytest tests/plugins/model_providers/test_zai_profile.py tests/agent/test_reasoning_effort_module.py -q` → 70 passed.
2. Red-first: revert the two production files and re-run — `test_graded_efforts_pass_through[medium-low]` and `test_disable_becomes_cheapest_level_not_a_disable_marker` both fail.
3. Live probes (2026-08-29, `glm-5.3-flash`). `1113` = insufficient balance, i.e. **the parameter passed validation**; `1210` = rejected:

   ```
   api.z.ai/api/paas/v4        low 1113 | medium 1210 | high 1113 | max 1113
                               xhigh 1210 | minimal 1210 | none 1210
                               thinking:{"type":"disabled"} 1210
                               thinking:{"type":"enabled"}  1113

   api.z.ai/api/coding/paas/v4 medium 200 | thinking:{"type":"disabled"} 200
   ```

   Reproduce one line with:

   ```bash
   curl -s https://api.z.ai/api/paas/v4/chat/completions \
     -H "Authorization: Bearer $GLM_API_KEY" -H "Content-Type: application/json" \
     -d '{"model":"glm-5.3-flash","messages":[{"role":"user","content":"say ok"}],"max_tokens":8,"reasoning_effort":"medium"}'
   ```

## Checklist — Code
- [x] Conventional Commits — `fix(zai): …`
- [x] Searched for existing PRs — none open on GLM-5.3 effort
- [x] Only changes related to this fix (3 files, +63/-17)
- [x] `ruff check .` passes; no new ruff diagnostics on the touched files
- [x] Repo-wide `pytest` failure set is byte-identical to pristine `main` (15 pre-existing failures in `video_gen`/`hindsight`/`a2a`, unrelated)
- [x] Tests added, and proven to fail without the production change
- [x] Tested on: macOS 15 (Python 3.12)

## Checklist — Documentation & Housekeeping
- [x] Docstrings updated (zai module, `_is_glm_5_3`, `GLM53_EFFORTS`)
- [x] `cli-config.yaml.example` — N/A (no config keys; the vocabulary is data in `agent/reasoning_effort.py`)
- [x] CONTRIBUTING / AGENTS — N/A
- [x] Cross-platform — N/A (pure request-mapping logic)
- [x] Tool descriptions — N/A

### DIFF
--- a/agent/reasoning_effort.py
+++ b/agent/reasoning_effort.py
@@ -117,11 +117,20 @@ KIMI_K3_OVERRIDES: dict[str, str] = {"medium": "high", "xhigh": "max"}
 GLM52_EFFORTS: tuple[str, ...] = ("high", "max")
 GLM52_OVERRIDES: dict[str, str] = {"xhigh": "max"}
 
-#: GLM-5.3 widens the knob to a graded low/medium/high/max scale — verified
-#: live on api.z.ai/api/coding/paas/v4 (issue #91789, 2026-08-21): every
-#: level accepted with monotonic reasoning-token scaling (low=4, medium=11,
-#: high=98, max=125 on the probe prompt). ``xhigh`` requests the top tier.
-GLM53_EFFORTS: tuple[str, ...] = ("low", "medium", "high", "max")
+#: GLM-5.3 widens GLM-5.2's knob downward to low, but the vocabulary is
+#: exactly ``low|high|max`` — the server enumerates it in its own rejection:
+#: ``1210: This model always engages in thinking and cannot be disabled;
+#: please use low, high, or max``. Probed live 2026-08-29 against
+#: api.z.ai/api/paas/v4 (the base_url the zai profile ships) on
+#: glm-5.3-flash: low/high/max pass validation, ``medium`` is rejected 1210
+#: alongside ``minimal``/``xhigh``/``none``. #91789 measured a graded
+#: low/medium/high/max scale on api.z.ai/api/**coding**/paas/v4, which is
+#: the more permissive route and does accept ``medium`` — but declaring
+#: ``medium`` here turned every request at that effort into a hard 1210 on
+#: the default route. The strict route's vocabulary is the safe one to
+#: declare: ``medium`` then clamps to ``low`` (nearest weaker) for coding-
+#: route users instead of failing outright for everyone else.
+GLM53_EFFORTS: tuple[str, ...] = ("low", "high", "max")
 GLM53_OVERRIDES: dict[str, str] = {"xhigh": "max"}
 
 #: DeepSeek V4 OpenAI-compat endpoint: low/medium/high/max; ``xhigh``
--- a/plugins/model-providers/zai/__init__.py
+++ b/plugins/model-providers/zai/__init__.py
@@ -22,6 +22,13 @@ two enabled levels — ``high`` and ``max`` — on the OpenAI-compatible endpoin
 (per Z.AI / BigModel docs).  Hermes' richer effort scale is collapsed onto
 those two so the user's effort preference actually reaches the model instead
 of being silently dropped.
+
+GLM-5.3 reaches one level lower (``low``/``high``/``max``) but drops the
+``thinking`` toggle: it always thinks, and rejects both a disable marker and
+any effort word outside that set with ``1210``.  So on 5.3 the toggle is
+never sent and an explicit disable resolves to the cheapest level rather
+than to nothing — omitting the field would leave the server default, which
+is the *most* expensive tier.
 """
 
 from __future__ import annotations
@@ -68,9 +75,11 @@ def _is_glm_5_2(model: str | None) -> bool:
 def _is_glm_5_3(model: str | None) -> bool:
     """Detect GLM-5.3 specifically — it has a wider effort vocabulary.
 
-    5.2 accepts only ``high``/``max``; 5.3 accepts a graded
-    ``low``/``medium``/``high``/``max`` scale (verified live, issue #91789),
-    so effort mapping must pick the vocabulary per model.
+    5.2 accepts only ``high``/``max``; 5.3 reaches one level lower with
+    ``low``/``high``/``max`` (the server enumerates that set in its own 1210
+    rejection), so effort mapping must pick the vocabulary per model. 5.3
+    also refuses to turn thinking off at all — see
+    :func:`_glm_5_2_reasoning_effort` and :meth:`ZaiProfile.build_api_kwargs_extras`.
     """
     m = (model or "").strip().lower()
     if not m:
@@ -93,7 +102,12 @@ def _glm_5_2_reasoning_effort(
     if not isinstance(reasoning_config, dict):
         return None
     if reasoning_config.get("enabled") is False:
-        return None
+        # GLM-5.3 has no off switch: the wire rejects a disable with
+        # ``1210 ... cannot be disabled``, and omitting the field leaves the
+        # server default (max — the most expensive tier), so "off" would cost
+        # *more* than any level the user could have asked for. The cheapest
+        # supported level is as close as the wire gets. 5.2 has no such trap.
+        return "low" if _is_glm_5_3(model) else None
 
     effort = (reasoning_config.get("effort") or "").strip().lower()
     if not effort or effort == "none":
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** Correct provider-side fix that wires missing `reasoning_effort` for GLM-5.3 family on Z.AI. Declares wire vocabulary as data (`low|high|max`) and maps internal effort ladder onto it, preserving `high→high`, `xhigh/ultra→max`, `none→low` (cheapest, not off), and omitting when no preference.

**What it does**
- `agent/reasoning_effort.py:31` adds `GLM53_EFFORTS=("low","high","max")` and `GLM53_OVERRIDES={"xhigh":"max","ultra":"max"}` with comment explaining always-on thinking (1210) and server-default `max` cost.
- `plugins/model-providers/zai/__init__.py:43` adds `_GLM_5_3_TOKENS=("glm-5.3","glm-5-3","glm-5p3")`, `_is_glm_5_3()`, `_glm_5_3_reasoning_effort()` (returns `low` for `none`, `None` when enabled false/empty, else `clamp_effort(...,GLM53_EFFORTS,GLM53_OVERRIDES)` fallback to `low`), and extends `build_api_kwargs_extras` guard to include `not _is_glm_5_3` and new branch that emits `top_level["reasoning_effort"]`.
- GLM-5.2 and GLM-4.5 unchanged (`xhigh→max` for 5.2, no emit for 4.5). Validated live: `reasoning_effort:max` on `glm-5.3-flash` now passes (1113 billing vs 1210 unsupported).
- Tests `tests/plugins/model_providers/test_zai_profile.py:105` cover `xhigh→max`, `high→high`, `medium→low`, `none→low`, no preference → omitted, 5.2 preserved, 4.5 untouched.

**Non-blocking notes**
- `enabled is False → None` (omit) means "user explicitly disabled thinking" keeps server default (`max`) for 5.3 — correct per note that off is rejected (1210). If user truly wants no reasoning, provider correctly ignores request.
- Token match is substring `in m.lower()` — `"glm-5.3-flash"` matches, but also `"my-glm-5.30"` would — acceptable; keep consistent with `_is_glm_5_2` pattern.
- No change to `thinking`/`extra_body` — GLM-5.3 thinking always on via separate field? Ensure not emitting `thinking: disabled` alongside `reasoning_effort`.

pfcpdh:
Independent reproduction + confirmation, via a different failure path than described above.

**Surface:** Hermes gateway (config hot-reload path). `agent.reasoning_effort: medium` in `config.yaml` → every gateway-threaded request to `glm-5.3-flash` failed with the 1210 error on the standard route, while a long-lived CLI process that had cached pre-change config succeeded — same credentials, same model, differing only in whether `reasoning_effort: medium` was emitted. Logs: `error_type=BadRequestError ... provider=zai ... HTTP 400: This model always engages in thinking and cannot be disabled; please use low, high, or max`.

**Live probe (2026-09-05, `glm-5.3-flash`, standard route):** `reasoning_effort: medium` → HTTP 400 code 1210; `reasoning_effort: high` → HTTP 200, correct response, `reasoning_tokens` present. Matches the PR's probe table exactly.

The fix (clamp `medium → low`, `GLM53_EFFORTS = ("low","high","max")`) behaves correctly in production after applying the equivalent change locally. This confirms the failure mode is not coding-route-specific and affects any hot-reloading client, not just fresh sessions. Strong support from me.

ivfarias:
There are now four open fixes for this same 1210 — #96849 (`medium`→`low`, profile only), #98465 (`medium`→`high`), #102764 (flash-scoped, plus a transport fix), and this one. Rather than argue for mine, I want to flag that two of them rest on incompatible diagnoses, and that's worth settling before any of them is picked.

**#102764 reads this as a model constraint** and gates on `is_glm53flash_model()`. **This PR reads it as a route constraint** and narrows the whole `glm-5.3*` vocabulary. My probes (2026-08-29) are what pushed me toward the second reading — same model, same key, same body, only `base_url` differing:

```
api.z.ai/api/paas/v4         low 1113 | medium 1210 | high 1113 | max 1113
                             xhigh 1210 | minimal 1210 | none 1210
                             thinking:{"type":"disabled"} 1210
                             thinking:{"type":"enabled"}  1113
api.z.ai/api/coding/paas/v4  medium  200 | thinking:{"type":"disabled"} 200
```

(1113 = insufficient balance, i.e. the parameter cleared validation. 1210 = rejected.)

This also explains how the wrong vocabulary got declared originally: #91789 measured `low/medium/high/max` against the coding route, which really does accept `medium`. The profile ships the standard route, which doesn't.

What that data does **not** establish is whether non-flash `glm-5.3` behaves the same, since every line above is `glm-5.3-flash`. If non-flash also 1210s on `medium`, the constraint is route-dependent and #102764's model gate leaves those users broken. If non-flash accepts `medium`, #102764 has the better diagnosis and this PR over-narrows. It's a single curl either way and I'm running it now — I'll post the table here regardless of which way it lands, and if it comes back flash-specific I'll close this in favour of #102764.

Separately, and independent of that question: #102764's `chat_completions.py` fix is real and this PR doesn't cover it. The transport defaults effort to `"medium"` when the user expressed no preference, so the 400 fires for people who never touched `/reasoning` at all. Whatever gate ends up being correct needs applying there too. Happy to fold that in here if consolidating into one PR is easier for review than merging two.

PR #97289, opened 2026-08-28
fix(bedrock): keep cachePoint out of toolConfig.tools for Nova models

### LINKED ISSUE #97281: [Bug]: Nova models reject cachePoint in toolConfig.tools - ValidationException
### Bug Description

When using Amazon Nova models (e.g., us.amazon.nova-pro-v1:0) via the Bedrock Converse API, Hermes inserts cachePoint markers into toolConfig.tools. This causes AWS Bedrock to reject the request with:

ValidationException: An error occurred (ValidationException) when calling the ConverseStream operation:
The model returned the following errors: Malformed input request: #/toolConfig/tools/18:
extraneous key [cachePoint] is not permitted, please reformat your input and try again.

Root Cause

Per-model cachePoint placement differs between Bedrock model families:

| Model family | cachePoint in system | cachePoint in toolConfig.tools |
|--------------|----------------------|--------------------------------|
| Nova         | ✅ accepted          | ❌ REJECTED server-side        |
| Anthropic    | ✅ accepted          | ✅ accepted                    |

The current code in agent/bedrock_adapter.py uses a single allowlist _CACHE_POINT_PATTERNS that includes both anthropic.claude and amazon.nova, and applies the same caching logic to both for tools. However, Nova models only support cachePoint in the system block and messages array — not in toolConfig.tools.

Reference: https://dev.to/rajmurugan/field-notes-turning-prompt-caching-on-for-a-production-bedrock-workload-2l8b

This issue was also encountered and fixed in the LibreChat project:
- https://github.com/danny-avila/LibreChat/issues/13838
- https://github.com/danny-avila/agents/pull/253

Impact

- Users trying to use Nova models with tools (any Hermes tool) get 100% failure rate
- Nova is listed as a supported model but is effectively broken when tools are enabled
- No config option exists to disable cachePoint insertion as a workaround

Workaround

None. There is no config option to disable cachePoint insertion for Bedrock.

### Steps to Reproduce

1. Configure Bedrock provider with valid AWS credentials
2. Switch to a Nova model: /model us.amazon.nova-pro-v1:0
3. Send any message that triggers a tool call
4. Observe ValidationException: extraneous key [cachePoint] is not permitted

### Expected Behavior

Calls to Bedrock using Nova model should suceed.

### Actual Behavior

An error occurred (ValidationException) when calling the ConverseStream operation: The model returned the following errors: Malformed input request: #/toolConfig/tools/4: extraneous key [cachePoint] is not permitted, please reformat your input and try again.

### Affected Component

Agent Core (conversation loop, context compression, memory)

N/A (CLI only)

### Debug Report

```shell
Debug report uploaded:
  Report     https://paste.rs/8ahC8
  agent.log  https://paste.rs/3V1vT
```

### Operating System

Debian 12

### Python Version

3.11.16 (in hermes venv)

### Hermes Version

Hermes Agent v0.20.6 (2026.8.27) · upstream 306db277

In `agent/bedrock_adapter.py`, lines ~1261-1300:
```python 
cache_enabled = _model_supports_prompt_cache(model)                                                                                                         

# ... later, in tools section:
if cache_enabled:
  converse_tools = converse_tools + [{"cachePoint": {"type": "default"}}]
```
The `cache_enabled` flag is True for both Nova and Anthropic models, but Nova cannot accept `cachePoint` in the tools array.

### Proposed Fix (optional)

_No response_

liuhao1024:
## What does this PR do?

Amazon Nova models on the Bedrock Converse API path reject any request where a `cachePoint` block appears inside `toolConfig.tools` — Bedrock fails the whole call with `ValidationException: #/toolConfig/tools/18: extraneous key [cachePoint] is not permitted`. Because `_CACHE_POINT_PATTERNS` treats `anthropic.claude` and `amazon.nova` identically, `build_converse_kwargs` appended the tool-level cache marker for Nova too, so any Nova model with tools enabled failed 100% of requests (#97281).

Nova does accept `cachePoint` in the system list and the messages array, so this PR splits capability detection by placement: a new `_model_supports_tool_cache_point()` (Anthropic-only allowlist) gates the `toolConfig.tools` marker, while `_model_supports_prompt_cache()` continues to gate the system and message markers. Claude behavior is unchanged; only the previously-broken Nova + tools combination changes.

## Related Issue

Fixes #97281

## Type of Change

## Changes Made

- `agent/bedrock_adapter.py`: added `_CACHE_POINT_TOOL_PATTERNS` + `_model_supports_tool_cache_point()`; the tool-level cachePoint append in `build_converse_kwargs` now uses this Anthropic-only gate instead of the shared `cache_enabled` flag.
- `tests/agent/test_bedrock_adapter.py`: added `test_nova_gets_system_and_message_cache_point_but_not_tools` (Nova keeps system + message markers, tools stay clean); updated the existing cachePoint test docstring, which previously claimed Nova also received the tools marker.

## How to Test

1. `uv run --extra dev python -m pytest tests/agent/test_bedrock_adapter.py -q`
   - Observed result: 67 passed (includes the new Nova regression test).
2. Fail-closed check for the new test: reverting the one-line gate back to `if cache_enabled:` makes `test_nova_gets_system_and_message_cache_point_but_not_tools` fail (tools list picks up the forbidden marker), confirming the test pins the bug.
3. Full bedrock surface: `uv run --extra dev python -m pytest tests/agent/test_bedrock_integration.py tests/agent/test_bedrock_empty_text_blocks.py tests/agent/test_bedrock_1m_context.py tests/hermes_cli/test_bedrock_model_picker.py tests/hermes_cli/test_bedrock_mantle_key_env.py tests/hermes_cli/test_bedrock_region_scoped_picker.py -q`
   - Observed result: 57 passed.

### Code

## For New Skills

N/A — not a skill.

## Screenshots / Logs

N/A — wire-format change verified by unit tests; the exact server-side error message is quoted in #97281.

### DIFF
--- a/agent/bedrock_adapter.py
+++ b/agent/bedrock_adapter.py
@@ -645,6 +645,21 @@ def _model_supports_prompt_cache(model_id: str) -> bool:
     return any(pattern in model_lower for pattern in _CACHE_POINT_PATTERNS)
 
 
+# Nova accepts cachePoint blocks in the system list and the messages array
+# but rejects them inside toolConfig.tools — Bedrock fails the whole request
+# with "extraneous key [cachePoint] is not permitted" (#97281), so tool-level
+# cache markers stay Anthropic-only.
+_CACHE_POINT_TOOL_PATTERNS = [
+    "anthropic.claude",
+]
+
+
+def _model_supports_tool_cache_point(model_id: str) -> bool:
+    """Return True if the model accepts a cachePoint block in toolConfig.tools."""
+    model_lower = model_id.lower()
+    return any(pattern in model_lower for pattern in _CACHE_POINT_TOOL_PATTERNS)
+
+
 def is_anthropic_bedrock_model(model_id: str) -> bool:
     """Return True if the model is an Anthropic Claude model on Bedrock.
 
@@ -1263,7 +1278,7 @@ def build_converse_kwargs(
             # Strip tools for known non-tool-calling models and warn the user.
             # Ref: PR #7920 feedback from @ptlally, pattern from PR #4346.
             if _model_supports_tool_use(model):
-                if cache_enabled:
+                if _model_supports_tool_cache_point(model):
                     converse_tools = converse_tools + [{"cachePoint": {"type": "default"}}]
                 kwargs["toolConfig"] = {"tools": converse_tools}
             else:
--- a/tests/agent/test_bedrock_adapter.py
+++ b/tests/agent/test_bedrock_adapter.py
@@ -504,8 +504,8 @@ class TestBuildConverseKwargs:
 
 
     def test_cache_point_added_for_supported_model(self):
-        """Claude and Nova on the Converse path get cachePoint markers on
-        system, tools, and the message before the newest turn."""
+        """Claude on the Converse path gets cachePoint markers on system, tools,
+        and the message before the newest turn."""
         from agent.bedrock_adapter import build_converse_kwargs
         tools = [{"type": "function", "function": {
             "name": "test", "description": "Test", "parameters": {},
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Minimal, correct Bedrock Converse fix that distinguishes prompt-cache vs tool-cache capabilities. Previously `cachePoint` in `toolConfig.tools` was sent for Nova and failed 100% of tool-enabled Nova calls (`extraneous key [cachePoint]`). Now gated Anthropic-only — Claude unchanged, Nova keeps system+message markers.

**What it does**
- `agent/bedrock_adapter.py:37` adds `_CACHE_POINT_TOOL_PATTERNS=["anthropic.claude"]` and `_model_supports_tool_cache_point()`; `build_converse_kwargs` changes `if cache_enabled:` → `if _model_supports_tool_cache_point(model)` for the `converse_tools + [{"cachePoint":...}]` append. System/message markers continue using `_model_supports_prompt_cache` (which includes `amazon.nova` via `_CACHE_POINT_PATTERNS`).
- Tests `tests/agent/test_bedrock_adapter.py:77` adds `test_nova_gets_system_and_message_cache_point_but_not_tools` asserting `system[-1]` is cachePoint, no tool equals cachePoint, `messages[-2]` is cachePoint for `us.amazon.nova-pro-v1:0`; prior test docstring updated from "Claude and Nova" to "Claude".

**Non-blocking notes**
- Allowlist is data-driven — future tool-capable models can extend `_CACHE_POINT_TOOL_PATTERNS` explicitly (good). Consider adding comment pointing to Bedrock docs for tool-level cachePoint support.
- Pattern match uses `in model_lower` substring — covers `us.amazon.nova-pro-v1:0`, `amazon.nova-micro`, etc. Verify no false positive on hypothetical `my-anthropic.claude-extended` — unlikely and safe.
- No prompt-caching cost regression — Nova still benefits from system/message cachePoints.

liuhao1024:
After independently reviewing #97327 against this fix, I'm stepping aside in its favor — it is the stronger repair on the axes that matter here:

- **Coverage.** The runtime Bedrock-verdict retry (parse the JSON pointer, drop the rejected placement, retry once, remember per model) heals *any* family whose placement rules drift, not just the Nova entry a static table happens to know about. My `_CACHE_POINT_TOOL_PATTERNS` allowlist only fixes the family someone has already triaged — the exact maintenance trap the module's own `_NON_TOOL_CALLING_PATTERNS` comment warns about.
- **Architecture fit.** The rejection-recovery flow mirrors the existing self-heal idiom in this module (`is_streaming_access_denied_error` → non-streaming `converse()` retry), so it extends an established pattern rather than adding a second static table.

One note the runtime approach may want to absorb: a static *initial* table for the already-known Nova case would keep the very first tool-enabled turn of each process from paying a doomed request + retry round-trip (the runtime path only learns after one failure). Seeding `_CACHE_POINT_REJECTIONS` with Nova's `tools` placement — or the equivalent gate in `build_converse_kwargs` — would combine both properties. Happy to see either shape land.

Closing this in favor of #97327. Thanks for the thorough work.

PR #94385, opened 2026-08-25
fix(cli): make Shift+letter actually type a capital under modifyOtherKeys

MwC-Trexx:
## Summary

- Patch prompt_toolkit's VT100 parser so character-valued `ANSI_SEQUENCES` entries insert the character, not the matched escape bytes
- Preserve intentionally blanked `insert_text` on multi-key ANSI tails (Alt / Shift+Alt must not type the letter)
- Add PromptSession-level regression tests for Shift+letter, Shift+Space, Alt, and Shift+Alt under modifyOtherKeys level 2 / CSI-u

## Motivation

#87511 mapped `ESC[27;2;<code>~` / `ESC[<code>;2u` to uppercase letters, but the classic CLI still leaks raw CSI into the buffer on Ghostty (and iTerm2 / WezTerm / kitty). prompt_toolkit builds `KeyPress(key=<table value>, data=<matched bytes>)` and the default `Keys.Any` binding inserts `event.data` — so mapping fixed what the key *is*, not what it *types*.

Verified on current `main` before this change: Shift+W then `hat` submits `\x1b[27;2;87~hat`. After: `What`.

## Changes

- `hermes_cli/pt_input_extras.py` — `_install_literal_key_data_patch()` (defensive getattr, marker on wrapper, only rewrites non-empty `insert_text != key`, called at end of `install_modify_other_keys_aliases()`)
- `tests/cli/test_modify_other_keys_insert_text.py` — asserts submitted text via real `PromptSession` + pipe input; includes Alt/Shift+Alt non-insert guards

## Related

- Fixes the remaining half of #87390 / #86866 after #87511
- Refs #87785 — same problem class (krunkosaurus); this is a rebased approach on current main that also keeps multi-key tail blanks intact. Happy to coordinate with that PR rather than race it.
- Orthogonal to #87637 (non-Latin Shift map extension) — that PR extends the table and inherits the same insert-text gap unless something like this lands

## Test plan

- [x] `./venv/bin/python -m pytest tests/cli/test_modify_other_keys_insert_text.py tests/cli/test_modify_other_keys_aliases.py -q` — **233 passed**
- [x] Pipe repro: `\x1b[27;2;87~hat \x1b[27;2;72~appens` → `What Happens`
- [x] Alt+a / Shift+Alt+a do not insert into the buffer
- [ ] Manual on Ghostty classic CLI: type capitals with Shift — no `[27;2;…~` leak; Shift+Enter still newline; Alt chords unchanged

## Notes for reviewers

[...]

PR #87785, opened 2026-08-16
fix(cli): make Shift+letter actually type a capital under modifyOtherKeys

krunkosaurus:
## What does this PR do?

Completes #87511. That PR mapped `ESC[27;2;<code>~` / `ESC[<code>;2u` to the uppercase character, but **Shift+letter still leaks its escape sequence into the prompt** — the mapping fixed what the key *is*, not what it *types*.

prompt_toolkit's VT100 parser reports every match as `KeyPress(key=<table value>, data=<matched bytes>)`, and the default `Keys.Any` binding inserts `event.data` — the bytes, not the key. For the hundreds of `Keys`-valued entries in `install_modify_other_keys_aliases()` that is invisible: bindings match on `key`, and `data` is never read. For the character-valued entries it is the whole ballgame:

```
ANSI_SEQUENCES["\x1b[27;2;72~"] = "H"
  -> KeyPress(key="H", data="\x1b[27;2;72~")
  -> insert_text("\x1b[27;2;72~")
```

**Verified against `main` at 7095e23eb2**, driving a real `PromptSession` over a pipe input:

| typed | submitted (main) | submitted (this PR) |
|---|---|---|
| `Hello World` | `\x1b[27;2;72~ello \x1b[27;2;87~orld` | `Hello World` |
| Ctrl+A then `X` on `bc` | `Xbc` | `Xbc` (unchanged) |

So the symptom in #87390 — capital letters printing `[27;2;<code>~` on Ghostty — is still live on `main` for anyone on an allowlisted terminal. Reported independently by a Ghostty 1.3.2 user this morning.

## Related Issue

Refs #87511, #87390, #86866, #87631, #87637.

Not a duplicate of #87637 (Cyrillic / all cased scripts): that PR extends the same mapping table and inherits the same gap — its assertions are also key-level, so it will report green while Cyrillic capitals still leak. This fix is orthogonal and makes both work.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `hermes_cli/pt_input_extras.py` — new `_install_literal_key_data_patch()`: narrows `KeyPress.data` to the character for character-valued table entries. Called at the end of `install_modify_other_keys_aliases()`, so it covers Shift+letter, Shift+Space and the keypad digits. Idempotent, guarded by `try`/`except` and a class-level flag, and a no-op if prompt_toolkit's internals move.
- `tests/cli/test_modify_other_keys_insert_text.py` — 14 tests.

[...]

PR #97290, opened 2026-08-28
fix(cli): register modified kitty keypad forms (follow-up to #87785)

### LINKED ISSUE #90640: [Bug]: CLI prints raw CSI-u sequences for all numeric keypad keys
### Bug Description

Windows classic CLI prints raw CSI-u sequences for all numeric keypad keys

After updating Hermes, all numeric keypad keys are broken in the classic CLI (hermes).

Numeric keypad numbers and operators are inserted as raw escape sequences instead of their characters. For example, Numpad / produces something similar to:

^[[57410u

Important isolation tests:

Normal keyboard keys work correctly.
All numeric keypad numbers/operators are affected.
PowerShell receives numeric keypad keys correctly.
scripts/keystroke_diagnostic.py receives them correctly; e.g. Numpad / reports key='/' data='/'.
hermes --tui handles the numeric keypad correctly.
Only the classic hermes CLI is affected.
The issue remains after updating to current origin/main.

Environment

Hermes Agent v0.20.4 (2026.8.18)
Windows 10
Native PowerShell
Python 3.11.15
Classic Hermes CLI (prompt_toolkit)
Current Hermes origin/main

This looks related to CSI-u handling in the classic CLI input path. Issue #78285 appears potentially related because it describes unmapped CSI-u sequences falling through in the prompt_toolkit CLI while Ink TUI handles them correctly.

### Steps to Reproduce

2. Start the classic Hermes CLI with `hermes` (or `hermes -p <profile>`).
3. At the Hermes input prompt, press any key on the numeric keypad, with Num Lock enabled.
4. Numeric keypad keys are inserted as raw escape/CSI-u sequences instead of the expected characters. For example, pressing Numpad `/` produces something similar to `^[[57410u`.
5. Exit the classic CLI and start Hermes with `hermes --tui`.
6. Press the same numeric keypad keys. They work correctly in the TUI.

Additional isolation test:
- Run `uv run python scripts\keystroke_diagnostic.py`.
- Numeric keypad keys are detected correctly there. For example, Numpad `/` reports `key='/' data='/'`.
- The regular (non-numpad) keyboard also works correctly.
- The issue affects both numeric keypad digits and operators (`/`, `*`, `-`, `+`, etc.).

### Expected Behavior

Numeric keypad keys should behave the same as their equivalent regular keyboard keys in the classic Hermes CLI.

For example:
- Numpad `/` should insert `/`
- Numpad `1` should insert `1`
- Numpad `+` should insert `+`

This is already the behavior observed in `hermes --tui` and in `scripts/keystroke_diagnostic.py`.

### Actual Behavior

In the classic Hermes CLI, numeric keypad keys are inserted as raw escape/CSI-u sequences instead of their expected characters.

For example, pressing Numpad `/` produces a sequence similar to:

^[[57410u

The issue affects all tested numeric keypad digits and operators.

The same keys work correctly outside the classic CLI:
- `hermes --tui`: works correctly
- `scripts/keystroke_diagnostic.py`: works correctly
- Regular keyboard keys: work correctly

Example diagnostic output for Numpad `/`:

key='/' data='/'

### Affected Component

CLI (interactive chat)

_No response_

### Debug Report

```shell
Report: https://paste.rs/MIhjo
agent.log: https://dpaste.com/B5APJM9VT
gateway.log: https://dpaste.com/EDSAXXYYA
gui.log: https://paste.rs/og5Yi
```

### Operating System

Windows 10

### Python Version

_No response_

### Hermes Version

_No response_

```shell

```

_No response_

### Proposed Fix (optional)

_No response_

despotak:
> **Correction (2026-08-28):** this PR originally said "Fixes #90640" and claimed no existing PR covered the keypad. Both were wrong, and I've rewritten the description below. @jackulau root-caused #90640 in [this comment](https://github.com/NousResearch/hermes-agent/issues/90640#issuecomment-) and showed that **#87785 already fixes the reported symptom**; I dismissed that PR by its title without reading the diff. What this PR covers is the narrower follow-up gap jackulau flagged at the end of that comment as *"not covered by anything currently open."* Apologies for the noise.

## What does this PR do?

Registers the **modified** forms of the kitty keypad block.

This is a follow-up to #87785 (and its sibling #94385), not a competing implementation. Per @jackulau's root-cause on #90640, `_install_literal_key_data_patch()` in #87785 fixes the *bare* keypad forms — `ESC[57410u` → `('/', '/')` — which is the symptom that issue reports. That analysis is correct and I verified it independently.

The gap it leaves, quoting that comment:

> Modified keypad presses are still unmapped in both PRs [...] The functional-key block only installs the bare `ESC[<cp>u` spelling, while every other block installs modifier variants via `_install_paired()`. [...] that is a small follow-up and **it is not covered by anything currently open**.

This is that follow-up.

Measured on this machine: **18/18 keypad keys unmapped** under Shift, Alt, Ctrl and Ctrl+Shift alike, while the bare forms behaved as jackulau describes.

### The fix does not invent meanings

Each modified keypad key **mirrors whatever its non-keypad equivalent already resolves to at that same modifier**:

- `KP_Left` reads `ESC[1;<mod>D`
- `KP_PageUp` reads `ESC[5;<mod>~`
- `KP_Enter` reads Enter's own `ESC[13;<mod>u`
- digit/operator keys read their character's CSI-u form

Those tables are already correct, so the keypad cannot drift away from the keys it stands in for. Where an equivalent is itself unmapped, the keypad key **stays unmapped** rather than acquiring behaviour its twin does not have. Ignore-mapped codes (locks, media keys, bare modifier events) stay consumed under every modifier instead of leaking.

### Relationship to #87785 — please sequence, don't merge blind

The two touch the same area and are complementary, but **the character-valued keypad keys need both**. Measured through the real parser:

| sequence | main | +#87785 | +#87785 +this |
|---|---|---|---|
| `ESC[57410u` (bare Numpad `/`) | `('/', '\x1b[57410u')` | `('/', '/')` | unchanged |
| `ESC[57410;2u` (Shift+Numpad `/`) | decomposes | decomposes | `('?', '?')` |
| `ESC[57417;2u` (Shift+KP_Left) | decomposes | decomposes | `Keys.ShiftLeft` |

Navigation keypad keys mirror `Keys`-valued twins and work from this PR alone; the digit/operator ones resolve to characters and therefore also want #87785's `data` narrowing to avoid re-introducing the same raw-`data` insert. **I'd suggest landing #87785 first.**

### One case still broken after both — answering jackulau's open question

That comment asked whether any terminal sends the explicit unmodified `;1` form, since it would still leak. It does still leak, with both patches applied:

```
ESC[57410;1u  ->  Escape, '[', '5', '7', '4', '1', '0', ';', '1', 'u'
```

I have not observed a terminal that emits it, so I have deliberately **not** patched it here — that would be adding a mapping for a case nobody has reported, which is the opposite of what the rest of this PR does. Flagging it as measured-and-still-open rather than fixing it speculatively.

## Related Issue

Follow-up to #87785, addressing the residual gap identified in #90640. **Not** `Fixes #90640` — #87785 closes that.

## Type of Change

## Changes Made

- `hermes_cli/pt_input_extras.py` — register modified forms for the kitty keypad PUA block, mirroring each key's non-keypad twin.
- `tests/cli/test_kitty_keypad_parity.py` — new file, 17 tests.

## How to Test

1. In a terminal speaking the kitty keyboard protocol, run `hermes`.
2. Press `Ctrl+KP_Enter`, `Shift+KP_Left`, `Alt+KP_Delete`, `Ctrl+Shift+KP_Up` (NumLock off).
3. Before: raw `ESC[57414;5u`-style text lands in the buffer. After: each behaves exactly like its arrow-cluster / Enter twin.
4. `./scripts/run_tests.sh tests/cli/test_kitty_keypad_parity.py` — 17 passed.

Verified: `Ctrl+KP_Left` → `ControlLeft`, `Shift+KP_Up` → `ShiftUp`, `Alt+KP_Delete` and `Ctrl+KP_Enter` both matching their twins exactly. 265 passed / 1 skipped across the five upstream key suites.

### Code

- [x] I searched for existing PRs — **corrected**: #87785 covers the bare-form symptom and I initially missed it by reading its title rather than its diff. This PR is scoped to the residual modified-key gap that PR leaves open.
- [x] I've run the test suite and all tests pass — `./scripts/run_tests.sh`: **17 passed, 0 failed**
- [x] I've tested on my platform: CachyOS (Linux 7.2.0-1-cachyos), Ghostty/Wayland, Python 3.11.16, Hermes v0.20.6 (2026.8.27)

- [x] Documentation — N/A
- [x] `cli-config.yaml.example` — N/A
- [x] `CONTRIBUTING.md` / `AGENTS.md` — N/A
- [x] Cross-platform impact considered — pure sequence-table registration, no platform branches; `scripts/check-windows-footguns.py` clean

### DIFF
--- a/hermes_cli/pt_input_extras.py
+++ b/hermes_cli/pt_input_extras.py
@@ -269,10 +269,54 @@ def _modify_other_keys_aliases(ANSI_SEQUENCES: dict, Keys) -> dict[str, object]:
             for mod in _lock_twins(m) if key is not None else ():
                 _put(twin_fmt.format(mod=mod), key)
 
-    for code, key_val in _kitty_functional_map(Keys).items():
+    functional_map = _kitty_functional_map(Keys)
+    for code, key_val in functional_map.items():
         _put(f"\x1b[{code}u", key_val)
         for mod in _lock_twins(1):  # with a lock on these arrive as ESC[<code>;129u etc.
             _put(f"\x1b[{code};{mod}u", key_val)
+
+    # -- Kitty functional keys WITH a real modifier (#90640) ----
+    # The block above covers the bare key and its lock twins (modifier 1), but kitty sends
+    # Ctrl+KP_Enter as ESC[57414;5u and Shift+KP_Left as ESC[57417;2u. Those have no entry at
+    # all, so every modified keypad press leaks its raw CSI into the buffer — kitty-only, since
+    # the PUA codepoints are its encoding alone. Rather than invent a meaning, mirror whatever
+    # the EQUIVALENT non-keypad key already resolves to at the same modifier: those tables are
+    # already correct, so KP_Left inherits exactly what Left does and cannot drift from it.
+    _KP_EQUIV_CSI = {57417: "D", 57418: "C", 57419: "A", 57420: "B",  # Left/Right/Up/Down
+                     57423: "H", 57424: "F"}                          # Home/End
+    _KP_EQUIV_TILDE = {57421: 5, 57422: 6, 57425: 2, 57426: 3}        # PgUp/PgDn/Insert/Delete
+    # KP_Enter resolves to Keys.ControlM rather than a character, so it matches neither branch;
+    # point it at Enter's own CSI-u codepoint so Ctrl+KP_Enter inherits the newline alias
+    # install_ctrl_enter_alias() registered before this builder runs.
+    _KP_EQUIV_CSIU = {57414: 13}
+    for code, key_val in functional_map.items():
+        for base_mod in (2, 3, 4, 5, 6, 7, 8):
+            for mod in _lock_variants(base_mod):
+                seq = f"\x1b[{code};{mod}u"
+                if code in _KP_EQUIV_CSI:
+                    source = f"\x1b[1;{mod}{_KP_EQUIV_CSI[code]}"
+                elif code in _KP_EQUIV_TILDE:
+                    source = f"\x1b[{_KP_EQUIV_TILDE[code]};{mod}~"
+                elif code in _KP_EQUIV_CSIU:
+                    source = f"\x1b[{_KP_EQUIV_CSIU[code]};{mod}u"
+                elif isinstance(key_val, str) and not isinstance(key_val, Keys) and len(key_val) == 1:
+                    source = f"\x1b[{ord(key_val)};{mod}u"
+                else:
+                    # Ignore-mapped codes (locks, media keys, bare modifier events) stay
+                    # consumed under every modifier instead of leaking.
+                    if key_val is Keys.Ignore:
+                        _put(seq, Keys.Ignore)
+                    continue
+                # Entries this builder already collected win over the installed table, because
+                # the modifier rows above are written into `aliases` and not yet into
+                # ANSI_SEQUENCES. Where the twin is itself unmapped, the keypad key stays
+                # unmapped too rather than acquiring behaviour its twin does not have.
+                equivalent = aliases.get(source)
+                if equivalent is None:
+                    equivalent = ANSI_SEQUENCES.get(source)
+                if equivalent is not None:
+                    _put(seq, equivalent)
+
     return aliases
 
 
--- a/tests/cli/test_kitty_keypad_parity.py
+++ b/tests/cli/test_kitty_keypad_parity.py
@@ -1,0 +1,44 @@
+"""Modified kitty keypad keys must mirror the keys they stand in for (#90640)."""
+
+from __future__ import annotations
+
+import pytest
+from prompt_toolkit.input.ansi_escape_sequences import ANSI_SEQUENCES
+
+from hermes_cli import pt_input_extras
+
+
+@pytest.fixture(autouse=True)
+def _aliases_installed():
+    pt_input_extras.install_shift_enter_alias()
+    pt_input_extras.install_ctrl_enter_alias()
+    pt_input_extras.install_modify_other_keys_aliases()
+
+
+# kitty PUA codepoint -> the sequence its non-keypad twin already resolves through.
+TWINS = {
+    57417: "\x1b[1;{mod}D",   # KP_Left   -> Left
+    57418: "\x1b[1;{mod}C",   # KP_Right  -> Right
+    57419: "\x1b[1;{mod}A",   # KP_Up     -> Up
+    57420: "\x1b[1;{mod}B",   # KP_Down   -> Down
+    57421: "\x1b[5;{mod}~",   # KP_PageUp -> PageUp
+    57426: "\x1b[3;{mod}~",   # KP_Delete -> Delete
+    57414: "\x1b[13;{mod}u",  # KP_Enter  -> Enter
+}
+
+
+@pytest.mark.parametrize("codepoint,template", sorted(TWINS.items()))
+@pytest.mark.parametrize("modifier", [2, 3, 5, 6])
+def test_modified_keypad_mirrors_its_non_keypad_twin(codepoint, template, modifier):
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Narrow follow-up that closes the modified-keypad gap left by #87785/#94385. Mirrors each PUA keypad codepoint's modified forms to whatever the non-keypad twin already resolves to at the same modifier, with lock-variant expansion — well-reasoned and not inventing semantics.

**What it does**
- `hermes_cli/pt_input_extras.py:43` adds `_KP_EQUIV_CSI` (57417→D etc. for arrows/home/end), `_KP_EQUIV_TILDE` (57421→5 etc. for PgUp/PgDn/Ins/Del) and `_KP_EQUIV_CSIU` (57414→13 for KP_Enter) and loops `for code, key_val in functional_map` × `base_mod in 2..8` × `_lock_variants(mod)` to install `\x1b[{code};{mod}u` → `ANSI_SEQUENCES.get(source)` where source mirrors the twin's existing sequence; `Keys.Ignore` stays consumed rather than leaking raw CSI.
- Drops `_IS_PREFIX_OF_LONGER_MATCH_CACHE` after installs (VT100 parser prefix cache invalidation) — needed as longer sequences flip prefix answers.
- Tests `tests/cli/test_kitty_keypad_parity.py:142` verify modified keypad mirrors twin for modifiers 2/3/5, bare forms still resolve, and lock-bit twins exist only for CSI-u (tilde must NOT gain lock twins).

**Non-blocking notes**
- `functional_map` iteration includes only codepoints already in `functional_map`; if a future kitty codepoint is added without native mapping, it won't get mirror — acceptable since tables would then have entry.
- `equivalent = ANSI_SEQUENCES.get(source)` skips when twin itself unmapped — correctly mirrors "where twin unmapped, keypad stays unmapped" rather than inventing.
- Large registration count (~18 keys × 7 mods × 4 lock variants) but still import-time cheap; no footprint on model/tool path.

despotak:
Rebased onto current `main` — this was `CONFLICTING` because the branch sat ~5994 commits behind and #102117 moved the code it patches. Same fix, re-derived; `MERGEABLE` again.

**The premise is now verifiable on `main` directly, which it was not when I opened this.** The description above says #87785 fixes the bare forms; since then #99488 merged and they resolve without any open PR applied. On clean `main` today:

```
bare  KP_/    ESC[57410u    -> '/'          # the symptom #90640 reports: fixed
bare  KP_Up   ESC[57419u    -> Keys.Up
Ctrl+ KP_Up   ESC[57419;5u  -> None         # the gap this PR closes: still live
Shift+KP_Left ESC[57417;2u  -> None
```

So the follow-up @jackulau flagged as *"not covered by anything currently open"* is still uncovered, and no longer depends on #87785 landing.

**What changed in the diff:** upstream turned `_modify_other_keys_aliases` into a builder returning an aliases dict rather than mutating `ANSI_SEQUENCES` in place, so the equivalent-lookup now reads the in-flight `aliases` first and the installed table second — the modifier rows it inherits from are written by that same builder and are not in `ANSI_SEQUENCES` yet. Tests cut from three to two invariants, proven red on base (revert `hermes_cli/pt_input_extras.py` to `main`, keep the test file, 29 of 31 fail). `scripts/run_tests.sh tests/cli/`: 1419 passed, 0 failed.

Nothing else changed; the design is unchanged from the description above.

teknium1:
Merged a slim rework of this fix in https://github.com/NousResearch/hermes-agent/pull/104869 at d8a07768c5ee59afae62e9fb51d45e1e30aa74da, with credit to @despotak. Modified keypad keys inherit their non-keypad mappings, including lock variants; the reported Alt+keypad Enter escape leak is verified fixed through the live CLI editor. Thank you for identifying and implementing the missing keypad coverage.

PR #94036, opened 2026-08-24
fix(compaction): preserve native capability across switches

steveonjava:
## What does this PR do?

Preserves native compaction capability when the runtime model changes. The switch path now derives capability from the effective endpoint when a same-provider direct endpoint is omitted, while provider-changing or unresolved endpoint switches remain default-deny. Rollback and fallback restoration preserve the prior capability state.

## Related Issue

Related prior art: #83272. This PR is independent: #83272 addresses a global trusted-origin allowlist and has no code dependency on this change.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `agent/agent_init.py`, `agent/agent_runtime_helpers.py`, `agent/chat_completion_helpers.py`, `agent/native_compaction.py`, `run_agent.py`: preserve and restore native compaction capability across runtime model switches, rollback, and fallback paths.
- `hermes_cli/model_switch.py`, `gateway/slash_commands.py`, `cli.py`, `tui_gateway/server.py`: resolve capability from the effective switch URL and keep ambiguous or provider-changing routes default-deny.
- `tests/run_agent/test_native_compaction_switch_capabilities.py`, `tests/run_agent/test_switch_model_context.py`, `tests/run_agent/test_switch_model_rollback.py`, `tests/run_agent/test_fallback_reasoning_override.py`: add regression coverage for direct endpoints, effective URLs, rollback, fallback restoration, and default-deny behavior.
- `tests/run_agent/test_switch_model_context.py`: update compatibility fixtures for the preserved runtime capability state.

## How to Test

1. Run the focused native-compaction regression tests. Result: 12 passed.
2. Run the expanded switch and fallback regression set. Result: 109 passed.
3. Run `ruff` on the touched Python files and `git diff --check`. Both passed.
4. CI completed successfully, including Linux, macOS, Windows, e2e, lint, and required-check jobs.

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
[...]

PR #97292, opened 2026-08-28
fix(gateway): preserve native compaction capability on resume

steveonjava:
> ## Reviewer action required before submission
> - The local record contains the targeted test run, not a local `pytest tests/ -q` run. Upstream CI passed its Python test job and platform jobs on Ubuntu, macOS, and Windows.
> - Confirm whether user-facing documentation or `cli-config.yaml.example` needs a change before requesting review.

## What does this PR do?

Preserves custom-provider native-compaction capability data when a gateway restores a persisted `/model` override after restart. The restored runtime now retains the requested provider, capability declarations, and output cap, so an explicitly trusted OpenAI-compatible proxy follows the intended compaction route without weakening the existing route gates.

## Related Issue

N/A. No upstream issue is linked for this restart-persistence defect.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- Normalize boolean custom-provider capabilities in the runtime bundle.
- Preserve `requested_provider`, `capabilities`, and `max_tokens` through resumed override resolution, turn construction, and gateway cache keys.
- Permit native compaction for an explicitly declared trusted proxy while retaining the existing direct-route checks.
- Add restart-persistence and cache coverage for the restored runtime.

## How to Test

1. Configure a named custom provider with `openai_native_compaction: true` and persist a `/model` override.
2. Simulate a gateway restart and verify the restored turn retains the requested provider, capability declaration, and output cap.
3. Run the targeted test command recorded in this PR. It completes with 211 passing tests. Upstream CI is green, including Python, macOS, and Windows jobs.

### Code

- [x] I've tested on my platform: Ubuntu, macOS, and Windows in upstream CI

- [x] I've updated `CONTRIBUTING.md` or `AGENTS.md` if I changed architecture or workflows — or N/A: no architecture or workflow change
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A: upstream CI passed macOS and Windows jobs
- [x] I've updated tool descriptions/schemas if I changed tool behavior — or N/A: no tool behavior changed

## Screenshots / Logs

Targeted tests: 211 passed. The required GitHub Actions checks, including Python tests and macOS and Windows jobs, are green.

### DIFF
--- a/agent/agent_init.py
+++ b/agent/agent_init.py
@@ -613,6 +613,7 @@ def init_agent(
     checkpoint_max_file_size_mb: int = 10,
     pass_session_id: bool = False,
     requested_provider: str = None,
+    capabilities: Optional[Dict[str, bool]] = None,
 ):
     """
     Initialize the AI Agent.
@@ -712,6 +713,10 @@ def init_agent(
         if isinstance(requested_provider, str) and requested_provider.strip()
         else agent.provider
     )
+    agent.capabilities = {
+        key: value for key, value in (capabilities or {}).items()
+        if isinstance(key, str) and isinstance(value, bool)
+    }
     agent._credential_pool = credential_pool
     agent.acp_command = acp_command or command
     agent.acp_args = list(acp_args or args or [])
--- a/agent/native_compaction.py
+++ b/agent/native_compaction.py
@@ -169,7 +169,10 @@ def native_compaction_context_management(
         return None
     if not is_native_compaction_model(getattr(agent, "model", None)):
         return None
-    if not is_direct_openai_route(
+    trusted_proxy = bool(
+        getattr(agent, "capabilities", {}).get("openai_native_compaction", False)
+    )
+    if not trusted_proxy and not is_direct_openai_route(
         getattr(agent, "base_url", None), is_codex_backend=is_codex_backend
     ):
         return None
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -3019,6 +3019,8 @@ def _resolve_runtime_agent_kwargs_for_provider(provider: str) -> dict:
         "command": runtime.get("command"),
         "args": list(runtime.get("args") or []),
         "credential_pool": runtime.get("credential_pool"),
+        "capabilities": dict(runtime.get("capabilities") or {}),
+        "max_tokens": runtime.get("max_output_tokens"),
     }
 
 
@@ -8215,11 +8217,13 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
             override_model = override.get("model", model)
             override_runtime = {
                 "provider": override.get("provider"),
+                "requested_provider": override.get("requested_provider"),
                 "api_key": override.get("api_key"),
                 "base_url": override.get("base_url"),
                 "api_mode": override.get("api_mode"),
                 "max_tokens": override.get("max_tokens"),
                 "credential_pool": override.get("credential_pool"),
+                "capabilities": dict(override.get("capabilities") or {}),
             }
             if override_runtime.get("api_key"):
                 if override_runtime.get("credential_pool") is None:
@@ -8368,6 +8372,7 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
             "args": list(runtime_kwargs.get("args") or []),
             "credential_pool": runtime_kwargs.get("credential_pool"),
             "max_tokens": runtime_kwargs.get("max_tokens"),
+            "capabilities": dict(runtime_kwargs.get("capabilities") or {}),
         }
         route = {
             "model": model,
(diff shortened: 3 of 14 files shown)


Enough1122:

The plumbing is clean: `capabilities` is sanitized at every boundary (string keys, bool values) in `agent_init.py`, `config.py` `_normalize_custom_provider_entry`, and `_lift_model_capabilities`, threaded through the runtime resolution chain and the session-override persistence path, and included in `_agent_config_signature` so a capability flip correctly busts the agent cache. The test coverage is unusually complete for this kind of pass-through change — signature change, rehydration across restart, turn-config propagation, and the native-compaction behavior test all assert the real end state.

A few things worth looking at:

1. **Inconsistent sanitization in `_get_named_custom_provider`** (`hermes_cli/runtime_provider.py`, the two `result["capabilities"] = dict(entry["capabilities"])` sites at ~L848-850 and ~L897-899). Everywhere else — `_lift_model_capabilities`, the config normalizer, and `agent_init` — the dict is filtered down to `str` keys / `bool` values, but these two sites copy the entry's `capabilities` verbatim. The `isinstance(..., dict)` guard prevents a crash, and the downstream `agent_init` filter happens to strip bad values later, so it's functionally safe today — but the two branches of the same function sanitize differently, and a future caller that reads `runtime["capabilities"]` before agent init (there are several in this very diff) will see the unfiltered version. Routing both sites through `_lift_model_capabilities` (or a shared `_filter_capabilities` helper) removes the divergence.

2. **`sorted((runtime.get("capabilities") or {}).items())` in `_agent_config_signature`** (`gateway/run.py` ~L26956) is a cache key — it will raise `TypeError` if any capability value is ever non-comparable (e.g. a dict smuggled past the sanitizers). Since the values are already bool-only upstream, this is currently safe, but keying on `frozenset(...items())` or sorting `(k, type(v), v)` makes the signature robust to a future sanitizer regression rather than throwing from inside the cache-busting path.

3. **Trust boundary deserves explicit documentation.** `openai_native_compaction` is the one capability that gates whether native-compaction payloads (which contain the conversation/context material) flow through a proxy that is *not* a direct OpenAI route. The name is generic enough that a user could flip it in config without appreciating the trust implication. One sentence in the config docs — "this sends native compaction payloads to the configured base_url; only set it for endpoints you trust with conversation content" — would make the risk legible at the point of configuration.

4. Minor: the `requested_provider` additions in `_resolve_session_agent_runtime` / `_apply_session_model_override` now persist a provider selector into the override and re-apply it on rehydrate — consistent with `api_mode`/`base_url` handling, but worth one test pinning that a *stale* `requested_provider` (provider removed from config since the session was saved) degrades gracefully rather than failing closed on rehydrate.

steveonjava:
Thanks for the review. I agree that the named-provider path should use the same string-key/boolean-value filter as the other runtime paths.

While tracing that path, I found a larger propagation gap: the normal gateway resolver did not forward capabilities at all. A configured trusted proxy therefore did not receive native compaction on the initial route. The rehydrated /model path did, which made behavior depend on a gateway restart. Commit `298af6190e94b191f7bf94537595c4c000d86ae1` fixes both paths, centralizes the filter, and adds regression coverage for an initial route and a resumed override.

That also resolves the cache-key concern. The signature now receives only filtered string/bool pairs, so it can keep its deterministic sorted representation without adding a second defensive encoding layer.

I added a configuration warning that enabling `openai_native_compaction` sends native-compaction conversation content to the configured `base_url` and should be limited to endpoints the operator trusts.

I do not think a stale `requested_provider` test applies here. The persisted override contains only model, provider, and base_url. `requested_provider` is re-derived from the live resolver during rehydration. The existing credential re-resolution failure behavior is the relevant fallback case.

Verification: `uv run --extra dev pytest -q tests/gateway/test_agent_cache.py tests/gateway/test_session_model_override_persistence.py tests/hermes_cli/test_runtime_provider_resolution.py tests/run_agent/test_native_compaction.py` (156 passed); `uv run --extra dev ruff check gateway/run.py hermes_cli/runtime_provider.py tests/gateway/test_agent_cache.py tests/hermes_cli/test_runtime_provider_resolution.py`.

steveonjava:
Follow-up: I found and fixed the live `/model` path too. `5c3c608cc4cb1a109ed37bb571e1493090f86092` carries normalized runtime capabilities through the switch result and in-memory session override, so the capability takes effect before restart. The focused suite now passes with 157 tests.

teknium1:
Merged via PR #98546 together with your #94036 (stacked salvage — this PR's runtime_capabilities plumbing builds on #94036's resolution machinery). All 3 commits cherry-picked with authorship preserved. Thank you @steveonjava!

steveonjava:
Glad this series of bugs is resolved and thanks for taking care of this @teknium1!

Issue #95681, opened 2026-08-26
Tracking: context-burden reduction campaign (COMPLETE) — ~15.2K tok/call saved common installs, ~20.2K desktop (−63%), up to ~17.7K messenger; 36 merged surfaces incl. core-tool deferral (#97979); skills curation (81→58, index −515); all 22 platform hints + every prompt block source-verified; capability truth layer (voice transcode, any-file MEDIA:, presence gates)


teknium1:
**STATUS: CLOSED 2026-09-02 — campaign complete.** 36 merged rows (Aug 25 → Sep 2, 2026). Final phase: #97979 core-tool deferral. Headline: **~15.2K tok/call saved on a common CLI install, ~20.2K on desktop-app sessions (~31.7K → ~11.7K, −63%), up to ~17.7K on messenger sessions** — plus compaction's verbatim tail 170K → 10–25K per compaction (#95571). Follow-ups live in #73179 (bridge arg validation) and #91125 (deferred-tool pins).

Tracking issue for the context-burden reduction campaign. Goal: one place to collect the numbers so we can produce a single show-off infographic covering the whole series.

## The series (merged)

| PR | surface | before | after | saved |
|---|---|---|---|---|
| #94599 | `bfl_flux3_*` promo tools removed (expired promo; + config migration v39 strips the `bfl` toolset) | ~2,260 tok/call | 0 | **−2,260/call** (every Nous-signed-in session) |
| #95005 | system prompt — Nous Subscription block removed | ~1,180 tok/call | 0 | **~1,180/call** (every call, every Nous-signed-in session) |
| #95287 | `cronjob` tool schema (was the single most expensive core tool) | 2,234 tok/call | 1,095 | **−1,139/call (−51%)** |
| #95384 | `computer_use` schema (2,237 → 2,064: max_elements removed, fat params dieted, ladder → response verdicts) + its ~1,232-tok system-prompt block DELETED (essentials folded into the schema) (+ honesty fix: background-FIRST pitch) | 3,469 tok/call (schema 2,237 + prompt 1,232) | 2,064 (schema only, prompt 0) | **−1,405/call (−41%)** |
| #95570 | `session_search` tool schema — teaching moved into response hints | 1,570 tok/call | 695 | **−875/call (−56%)** |
| #95571 | compaction verbatim tail — legacy hoard (170K kept on a 1M window) now opt-in | up to ~170K kept | 10–25K kept | **up to ~145–160K per compaction** on big-window sessions |
[...]

PR #97045, opened 2026-08-28
fix(curator): refuse un-ledgered terminal writes to the skills tree (#96962)

JoaoMarcos44:
## Summary

`hermes curator rollback` replays the whole-directory before/after snapshots that `tools/skill_ledger.py` captures on every `skill_manage` mutation. A `mv`/`cp`/`rm` issued through the **terminal** tool bypasses that capture entirely — the package changes on disk and the ledger never learns about it.

The curator prompt actively steered the fork into that blind spot. The "demote to references/templates/scripts" step read:

> Then archive the old sibling. Use `terminal` with `mkdir -p ~/.hermes/skills/<umbrella>/references/ && mv ... <umbrella>/references/<topic>.md`

So support files were re-homed with **no ledger entry**, and the archive that followed faithfully snapshotted an already-stripped package (`files: 1`). Rollback then restored a hollow skill — `SKILL.md` back, `references/` gone — recoverable only by hand out of the pre-run `.curator_backups` tar. That is the incident in #96962, and it matches the reporter's own root-cause correction on the issue: *"terminal-based mutations of the skills tree by background forks are un-ledgered and un-rollback-able."*

This PR closes the hole at both ends.

## Flow

```mermaid
%%{init: {'theme': 'dark', 'themeVariables': { 'primaryColor': '#8b0000', 'mainBkg': '#0a0204', 'primaryTextColor': '#ffccd5', 'primaryBorderColor': '#ff0038', 'lineColor': '#ff0038'}}}%%
graph TD
    A[Curator Fork - origin background_review] --> B{Re-home Support Files}

    B -->|BEFORE - terminal mv| C[Raw Shell Write]
    C --> D[Package Stripped On Disk]
    D --> E[No Ledger Entry]
    E --> F[Archive Snapshots files 1]
    F --> G[Rollback Restores Hollow Skill]

    B -->|AFTER - ledgered pair| H[skill_manage write_file]
    H --> I[skill_manage remove_file]
    I --> J[skill_manage delete absorbed_into]
    J --> K[record_mutation whole-dir snapshot]
    K --> L[Rollback Restores Full Package]

    C -.->|NEW - refused pre-approval| M[skills_write_guard]
    M --> H
```

## Changes

**`agent/curator.py` — remove the instruction that caused it**

- The re-homing step now prescribes the ledgered pair: `skill_manage write_file` on the umbrella → `remove_file` on the source → `delete` to archive.
[...]

PR #96980, opened 2026-08-28
fix(curator): restore full skill packages on ledger rollback (#96962)

HexLab98:
## Summary
- Curator consolidation re-homes `references/` / `scripts/` into an umbrella before `skill_manage(delete)`, so the ledger `delete` entry captured only the leftover `SKILL.md` (`files: 1`). `hermes curator rollback <entry-id>` then restored a hollow skill.
- Delete, archive, and purge snapshots now union in files from the newest curator `skills.tar.gz` for that package. Disk hashes win; missing support files are filled from the pre-run snapshot.
- Rollback of those actions does the same fill, so already-recorded incomplete entries can still recover the full package.

Fixes #96962

## Test plan
- [x] `scripts/run_tests.sh tests/tools/test_skill_ledger.py tests/hermes_cli/test_curator_archive_prune.py tests/agent/test_curator_backup.py`
- [x] `scripts/run_tests.sh tests/tools/test_skill_manager_tool.py tests/hermes_cli/test_curator_run.py`
- [ ] Create a skill with `references/extra.md`, snapshot, re-home/remove the support file, delete, then `hermes curator rollback <delete-id>` and confirm `references/extra.md` is restored.

Issue #96962, opened 2026-08-28
[Bug]: curator rollback restores SKILL.md but not support files (references/scripts) — rollback is not a full restore


zengzheqing:
## Bug Description

When the curator consolidates skill A into an umbrella, the ledger records a `delete` entry for A — but the entry captures only A's `SKILL.md` (`files: 1` in the ledger record), not A's `references/` / `scripts/` / `assets/`. A later `hermes curator rollback <delete-entry-id>` therefore restores a hollow skill: the SKILL.md is back, but every support file is gone. In our incident the absorbed skill's `references/pi-lesson-roadmap.md` survived only inside the umbrella (where it had been re-homed) and in the pre-run `.curator_backups` tar — neither is part of the rollback path.

## Reproduction

1. Create a skill with `references/extra.md`.
2. Let a curator consolidate pass absorb it (or `hermes curator archive` + delete flow).
3. `hermes curator rollback <entry-id>`.
4. The skill dir returns with SKILL.md only; `references/` is missing.

## Expected behavior

Ledger `delete`/`archive` entries capture the full skill package (or rollback restores from the pre-run tar snapshot), so rollback is a true restore.

## Evidence

- Ledger entry `f2d86faed30e` (action: delete): `files: 1`, while the skill on disk had a `references/` subdir.
- Recovery required manually extracting the file from `.curator_backups/<timestamp>/skills.tar.gz`.

Rollback fidelity matters because rollback is the documented recovery path for bad consolidations — a partial restore silently loses content.

zengzheqing:
**Correction after root-cause verification** — my initial framing mis-attributed the mechanism. Reading the ledger entries directly:

- The archive entry's before-state captured exactly one file (SKILL.md) because **at archive time the original dir had already been stripped** — the support files had been moved into the umbrella earlier in the pass.
- Those re-homing moves produced **no ledger entries at all**: the curator fork had worked around the (then-broken) skill_manage patch path by using terminal `mv`/`cp`, which bypasses ledger capture entirely. (In the same run's report the fork notes it used terminal edits for agent-created skills because the read-before-write guard refused patches.)

So the ledger's capture itself is faithful (whole-dir snapshots); the hole is that **terminal-based mutations of the skills tree by background forks are un-ledgered and un-rollback-able**. The fix for the underlying bypass is #95449 (read-before-write marks survive tool-worker snapshots), which restores the skill_manage path — with it, deletes/archives go through `record_mutation` and capture the full package again.

[...]

PR #97295, opened 2026-08-28
feat(skill_manage): operations[] is the call — each op names its skill; atomic with cross-skill rollback

teknium1:
Maintainer-approved follow-up to #97152, with TWO interface corrections during review (final shape is the maintainer's):

**`skill_manage(operations=[{name, action, ...}, ...])`** — the call IS the operations array. Each op names its target skill; a single edit is a list of one; a batch is a list of several, and ops may span DIFFERENT skills.

## Semantics
- **Atomic across everything touched**: every skill named in the batch is snapshotted before any op runs; any failure rolls ALL touched skills back (batch-created skills are removed). `failed_index` + the failing op's own teaching error return.
- **Ops**: create (must precede that skill's other ops), patch (old_string/new_string; content = full rewrite), write_file/remove_file, delete (sole op of its call — routes to the real delete handler; absorbed_into/recoverable-archive semantics untouched).
- **Same-file clobber guard** (maintainer concern, live-repro'd first): under last-wins sequencing, a double write to one path, write-then-remove, or a full rewrite after an earlier SKILL.md edit SILENTLY discards work — all rejected up front with teaching errors, before any side effect. Patch chains on one file (each op building on the last) stay legal, per-skill scoped.
- **Approval gate composes**: gated batches stage as ONE pending record (gist names ops + skills); `apply_skill_pending` replays the whole batch.
- Ops execute through the single-op path — ledger, telemetry, prompt-cache invalidation, sync-push fire per-op as today. Legacy flat shape still handler-accepted (old transcripts, staged replay), unadvertised.

## Schema cost (#95681 accounting)
Diet base 427 → **482 (+55/call)** for batch + cross-skill + rollback. (Review history: flat+optional-batch draft measured 577; the maintainer's single-shape correction bought back ~95.)

## Verification
Live E2E of the exact maintainer shape: single-as-list ✅ · multi-op one skill ✅ · cross-skill batch in one call ✅ · sole-op delete ✅ · cross-skill failure → alpha's patch undone AND batch-created beta removed ✅ · clobber guards fire pre-effect, chains legal ✅ · legacy flat accepted ✅ · gate stages one record, replay applies ✅. 58 passed; 1 pre-existing symlink failure (Windows env class, clean-main verified).

### DIFF
--- a/tests/tools/test_skill_manage_batch.py
+++ b/tests/tools/test_skill_manage_batch.py
@@ -36,7 +36,11 @@ class TestSkillManageBatch(unittest.TestCase):
         shutil.rmtree(self.home, ignore_errors=True)
 
     def _call(self, name, ops):
-        return json.loads(self.smt.skill_manage(action="", name=name, operations=ops))
+        # Inject the per-op name (tests were written per-skill; the
+        # interface is name-per-op, maintainer-directed).
+        for op in ops:
+            op.setdefault("name", name)
+        return json.loads(self.smt.skill_manage(action="", name="", operations=ops))
 
     def test_create_plus_files_atomic(self):
         r = self._call("probe", [
@@ -130,6 +134,25 @@ class TestSkillManageBatch(unittest.TestCase):
             {"action": "patch", "old_string": "H.", "new_string": "I."},
         ])
         self.assertTrue(r["success"], r)
+
+    def test_cross_skill_batch_and_rollback(self):
+        """Ops may target DIFFERENT skills; a late failure rolls back
+        every touched skill, including removing a batch-created one."""
+        self._call("alpha", [{"action": "create", "content": SK.format(n="alpha")}])
+        r = json.loads(self.smt.skill_manage(action="", name="", operations=[
+            {"name": "alpha", "action": "patch",
+             "old_string": "Step 1.", "new_string": "Step A."},
+            {"name": "beta", "action": "create", "content": SK.format(n="beta")},
+            {"name": "beta", "action": "write_file",
+             "file_path": "bad/nope.md", "file_content": "x"},
+        ]))
+        self.assertFalse(r["success"])
+        self.assertEqual(r["failed_index"], 2)
+        # alpha's patch undone; beta (batch-created) removed entirely.
+        content = open(os.path.join(self.home, "skills", "alpha", "SKILL.md")).read()
+        self.assertIn("Step 1.", content)
+        self.assertNotIn("Step A.", content)
+        self.assertFalse(os.path.exists(os.path.join(self.home, "skills", "beta")))
 
     def test_single_op_path_unchanged(self):
         self._call("probe", [{"action": "create", "content": SK.format(n="probe")}])
--- a/tests/tools/test_skill_manage_batch.py
+++ b/tests/tools/test_skill_manage_batch.py
@@ -71,10 +71,19 @@ class TestSkillManageBatch(unittest.TestCase):
         self.assertFalse(os.path.exists(os.path.join(self.home, "skills", "fresh")))
 
     def test_validation_rules(self):
-        # delete not batchable
+        # delete as SOLE op routes to the real delete (works)
+        self._call("probe", [{"action": "create", "content": SK.format(n="probe")}])
         r = self._call("probe", [{"action": "delete"}])
+        self.assertTrue(r["success"], r)
+        self.assertFalse(os.path.exists(os.path.join(self.home, "skills", "probe")))
+        # delete mixed with other ops rejected
+        self._call("probe", [{"action": "create", "content": SK.format(n="probe")}])
+        r = self._call("probe", [
+            {"action": "patch", "old_string": "Step 1.", "new_string": "X."},
+            {"action": "delete"},
+        ])
         self.assertFalse(r["success"])
-        self.assertIn("standalone", r["error"])
+        self.assertIn("SOLE", r["error"])
         # create must be first
         r = self._call("x", [
             {"action": "write_file", "file_path": "references/a.md", "file_content": "a"},
--- a/tests/tools/test_skill_manage_batch.py
+++ b/tests/tools/test_skill_manage_batch.py
@@ -97,6 +97,40 @@ class TestSkillManageBatch(unittest.TestCase):
         self.assertFalse(r["success"])
         self.assertIn("capped", r["error"])
 
+    def test_intra_batch_conflict_guard(self):
+        """Same-file double writes and post-edit full rewrites are always
+        a confused plan under last-wins sequencing — rejected BEFORE any
+        side effect. Patch chains and rewrite-first stay legal."""
+        self._call("probe", [{"action": "create", "content": SK.format(n="probe")}])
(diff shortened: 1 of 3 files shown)


andrexibiza (commented):
Reviewed exact head `e7fa4678045c1d3c6efe6a21fa3ef960fcc2a74c` against `main@baa344dee76993f0444c18fc59a69738ccb339d0` (the PR commit is a direct child of that main object). I read the full two-file diff, the current `skill_manage` mutation/ledger paths, the new tests, the existing review surface, and the adjacent skill-write graph. Exact-head CI is green: CI `33191412909`, Docker `33191411951`, and Nix `33191411963`. There is one commit, so the per-commit hosted-green gate is satisfied.

The interface itself is a useful capability: one pending approval record for a coherent multi-file skill edit is much better than approval spam, excluding `delete` from the batch is a disciplined scope decision, and the basic rollback/create-failure tests cover the happy transaction shape well. I found two correctness blockers in the transaction boundary, plus one hard placement gate before I think the word **atomic** is true end-to-end.

### P1 — rollback has no ownership/lease and can erase an independent writer

In `tools/skill_manager_tool.py::_skill_manage_batch()` (new block beginning around head line 1529), the transaction is:

1. `_find_skill(name)` and `shutil.copytree(pre_dir, snapshot)`;
2. run each operation through the normal `skill_manage()` path;
3. if a later op fails, `_rollback()` re-resolves the name, `shutil.rmtree(post_dir)`, then copies the old snapshot back.

Nothing holds one physical-target authority from **snapshot → all child mutations → commit/rollback**, and rollback does not verify that the live directory still represents this batch's generation. A concrete interleaving is therefore legal today:

`batch snapshots S → op[0] produces B → another Hermes process/Desktop writer produces C → op[1] fails → batch rmtree(C) and restores S`.

That turns a local batch failure into loss of a successful independent mutation. The failed-create variant is even sharper: when no preexisting skill existed, rollback removes whichever same-named skill `_find_skill()` sees at rollback time.

This is the same writer-serialization defect class already worked by `Guerilla-ops` in #75434, which explicitly identifies the lookup-to-write interval and rollback-over-in-flight-write hazard. `mooserini`'s #96059 is complementary on the staged-approval side: it carries a target precondition and cross-process lease through approved replay so stale authority cannot overwrite newer state. Neither PR is a duplicate of this batch API, but this PR needs to compose with that ownership model (or an equivalent current-main owner), not create a third transaction authority.

**Required fix:** acquire one target-scoped lease/generation before snapshot and hold it through final settlement, including rollback; rollback must refuse/reconcile rather than overwrite if its ownership proof is no longer current. Please add an adversarial witness that pauses after op 0, attempts an independent write from another process/session, then forces op 1 to fail and proves no independent bytes are destroyed. Cover the initially-absent/same-name-create case too.

### P1 — filesystem rollback does not roll back the durable settlement record

The PR body is explicit that each child op uses the normal single-op path so ledger entries and telemetry fire per op. That is precisely where the all-or-nothing claim breaks.

On every successful child op, current `skill_manage()` captures and appends a `tools.skill_ledger.record_mutation(...)` entry with before/after blobs, then mutates usage telemetry (`record_created` / `bump_patch`) before the batch knows whether a later child will fail. If op 1 then fails, `_rollback()` only restores the directory bytes. It does **not** invalidate/compensate the already-successful ledger entries or usage state.

The result is a durable audit trail that says a mutation survived even though the batch erased it. Worse, those entries remain individually actionable through `hermes curator rollback <entry-id>` because `skill_ledger.rollback_entry()` treats each recorded mutation as settled history. A failed create batch can likewise leave `record_created` provenance/usage state for a skill directory the batch removed.

This is the inverse side of the rollback-truth class exposed by #96962 and the complementary repairs #97045 (`JoaoMarcos44`) / #96980 (`HexLab98`): there, filesystem mutation escaped the ledger; here, the ledger can preserve effects that the filesystem transaction later revoked. Both break the same invariant: observable/auditable state must describe the state that actually settled.

**Required fix:** make side-effect publication transaction-aware. The cleanest shape is one batch transaction/receipt whose child effects remain provisional until commit; alternatively append explicit compensation/aborted records that make the final state and replay semantics unambiguous. A failed batch should leave filesystem + usage/provenance + rollback/audit view equivalent to the pre-batch state, except for a truthful `aborted`/`rolled_back` receipt if you want that history retained. Add a regression that inspects `skill_ledger.list_entries()` and usage state after a mid-batch failure, and prove none of the rolled-back child entries can later resurrect/revert state as if they had committed.

Prompt-cache invalidation is harmless to repeat, and the debounced sync hook can remain best-effort if it is proven to observe only the final settled state; I would not block on those independently. The durable ledger/provenance mismatch is the blocker.

### Hard landing gate — this commit regrows `tools/skill_manager_tool.py` past the 2K boundary

The final registration hunk is at head line ~2060 (`@@ -1867,6 +2060,7 @@`), so this one commit takes the production file back over 2,000 lines. The batch coordinator is also a genuinely separate responsibility: transaction admission, snapshot/rollback, child execution, and settlement do not belong inline in the already-central dispatcher.

Please extract the batch transaction owner into a bounded module and leave `skill_manager_tool.py` as the thin public/schema adapter. That also gives #75434/#96059-style lease and settlement semantics one place to compose rather than adding more authority to the godfile.

### Graph / credit / merge-order notes

- #97152 (`teknium1`) is the merged direct predecessor and explicitly sets up this `operations[]` follow-up; this PR correctly preserves that lineage.
- #95681 (`teknium1`) is the context-burden campaign umbrella; the +156-token schema spend here is a real capability cost, not a regression in #97152's diet, and the PR accounts for it honestly.
- #75434 (`Guerilla-ops`) is complementary prior work on cross-process skill-writer serialization/rollback ownership.
- #96059 (`mooserini`) is complementary prior work on exact-target preconditions and approved-replay leases.
- #97045 (`JoaoMarcos44`) and #96980 (`HexLab98`) are complementary rollback/ledger integrity work around #96962.
- #95449 (`zengzheqing`) independently changes the same `skill_manager_tool.py` background-review boundary and must survive any restack.

I would make those relationships explicit in the PR before landing so the eventual transaction owner preserves credit and the merge order is deliberate rather than file-conflict-driven. I am not suggesting merging any currently stale/conflicting branch wholesale; preserve the proven semantics and attribution while converging on one owner.

This is good, focused product work. Once the batch owns **one generation, one settlement, and one bounded implementation owner**, the API becomes substantially stronger than a filesystem undo wrapper: it becomes an actual transaction boundary. 🚀

teknium1:
Post-review aggressive live torture (39-scenario harness against a real isolated HERMES_HOME, plus registry-handler dispatch): found and fixed 3 holes in the same-file clobber guard in 12c27849da — patch-then-write and patch-then-remove on one supporting file silently discarded the patch, and './references/x.md' / 'references//x.md' path spellings bypassed the duplicate-write check. Guard is now one rule: any destructive op (write_file/remove_file/full rewrite) on a (skill, normalized-path) an earlier op touched is rejected pre-effect; additive patches stay legal (patch chains + write-then-patch). All 39 torture scenarios + 62 unit tests green.

andrexibiza:
🔥

PR #97298, opened 2026-08-28
fix(process): report recovered supervision loss honestly

### LINKED ISSUE #97285: Background-process supervisor reports false 'exited' (exit_code=null) for live children after backend restart; orphans keep running unsupervised
## Summary

The background-process supervisor (processes started via `terminal(background=true)`, managed with the `process` tool) reports **false "process exited" notifications** for children that are still running, after the desktop app backend restarts. The inverse (stale "alive" entries for dead processes) was also observed.

## Environment

- macOS 26.4.1 (Apple Silicon)
- Hermes desktop app (backend restart triggered by an app update, and by crashes)
- Long-running background children: headless CLI coding agents (60+ min runtimes), started with `background=true, notify=true`

## Observed behavior (2026-08-28, 4+ occurrences in one day)

1. Two long-running background processes drew "process exited" completion notifications **mid-run**. Both were verifiably alive at notice time (log/output-file mtimes kept growing; `pgrep` showed the pids). One later finished on its own and wrote its complete 33KB output file.
2. After an app-backend restart, a third background process showed `status: "exited"` with **`exit_code: null`** in `process(action="poll")` while the OS process kept running and writing to disk for ~30 more minutes (its git commits landed *after* the monitor declared it dead).
3. Separately, a stale entry reported a process as alive when the OS no longer ran it.

## Why it matters

- False completion notices invite duplicate dispatch: an operator (or agent) who trusts the notification relaunches the job, and the duplicate races the still-running original in the same workspace.
- The orphaned child keeps running **unsupervised** — for agentic children (coding agents with write access to a worktree) this is a real safety problem, not just a cosmetic one. In our case a full build's output had to be discarded because ~30 minutes of it ran with no supervision handle.

## Root cause (from the outside)

The supervisor appears to hold its child handles in the backend process. When the backend restarts, handles are lost; on reconnect the supervisor cannot re-attach, and reports the children "exited" (with `exit_code: null` — the tell that no real exit status was ever collected) instead of either re-adopting or killing them.

## Suggested fix

On backend restart/reconnect, for each child in the persisted process table:

1. Check whether the recorded pid (with start-time verification to avoid pid reuse) is still alive.
2. If alive: **re-adopt** it (re-attach output tailing where possible, or at minimum keep status `running` with a `reattached: true` flag), or — if re-adoption is impossible — **positively terminate it** so no orphan keeps writing.
3. Never emit a completion notification or `status: "exited"` unless a real exit status was collected; `exit_code: null` on an "exited" report should be treated as a bug invariant.
4. Prune table entries whose pids are gone (fixes the stale-alive case).

## Workarounds we use meanwhile

- Treat `status: "exited"` + `exit_code: null` as "handle lost", not "process dead".
- Verify liveness via `pgrep` + output-file mtime growth before trusting any exit notice.
- Policy: on lost supervision, kill the orphan first and judge its committed work afterwards.

fangliquanflq:
## Summary

- classify recovered process handles whose verified PID disappears as supervision loss rather than a collected exit
- return `status: "lost"` consistently from process poll/log/wait/list surfaces
- warn gateway users that supervision was lost without presenting `exit_code: null` as a successful completion

## Root cause

Checkpoint recovery correctly re-adopts a live host PID using its recorded start time, but when that detached PID later disappeared `_refresh_detached_session()` only set `exited = True` and left the default completion reason as `exited`. Query paths then derived `status: "exited"` from that boolean, while concise gateway notifications treated a missing exit code as success. No waitable process handle existed to justify either claim.

## Testing

- `scripts/run_tests.sh tests/tools/test_process_registry.py -k 'TestCheckpoint or TestGetAndPoll or TestListSessions' -q` (10 passed)
- `scripts/run_tests.sh tests/gateway/test_background_process_notifications.py -q` (24 passed)
- red-to-green regression: recovered PID disappearance previously returned `status: "exited"`; it now returns `status: "lost"` with `completion_reason: "lost"`

Closes #97285

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -3897,6 +3897,7 @@ def _format_concise_process_notification(
     exit_code,
     output: str,
     duration_seconds=None,
+    completion_reason: str = "exited",
 ) -> str:
     """One-line "pretty" completion message for the ``concise`` display mode.
 
@@ -3904,9 +3905,14 @@ def _format_concise_process_notification(
     the user can see what went wrong without the full raw dump. The full
     output always remains available to the agent via process(log/wait).
     """
-    ok = exit_code in {0, None}
-    icon = "✅" if ok else "❌"
-    verb = "finished" if ok else f"failed (exit {exit_code})"
+    lost = completion_reason == "lost"
+    ok = not lost and exit_code == 0
+    icon = "⚠️" if lost else ("✅" if ok else "❌")
+    verb = (
+        "supervision lost"
+        if lost
+        else ("finished" if ok else f"failed (exit {exit_code})")
+    )
     parts = [f"{icon} Background task {verb}"]
     short_cmd = _shorten_command_for_display(command)
     if short_cmd:
@@ -26686,7 +26692,8 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                     break
                 # Decide whether to notify based on mode
                 should_notify = (
-                    notify_mode in {"concise", "all", "result"}
+                    getattr(session, "completion_reason", "exited") == "lost"
+                    or notify_mode in {"concise", "all", "result"}
                     or (notify_mode == "error" and session.exit_code not in {0, None})
                 )
                 if should_notify:
@@ -26715,12 +26722,22 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                             session.exit_code,
                             new_output,
                             duration_seconds=_dur,
+                            completion_reason=getattr(
+                                session, "completion_reason", "exited"
+                            ),
                         )
                     else:
-                        message_text = (
-                            f"[Background process {session_id} finished with exit code {session.exit_code}~ "
-                            f"Here's the final output:\n{new_output}]"
-                        )
+                        if getattr(session, "completion_reason", "exited") == "lost":
+                            message_text = (
+                                f"[Background process {session_id} supervision was lost; "
+                                "no exit status was collected. Verify OS liveness before "
+                                f"restarting it. Last captured output:\n{new_output}]"
+                            )
+                        else:
+                            message_text = (
+                                f"[Background process {session_id} finished with exit code {session.exit_code}~ "
+                                f"Here's the final output:\n{new_output}]"
+                            )
                     adapter = None
                     for p, a in self.adapters.items():
                         if p.value == platform_name:
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -26204,8 +26204,11 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                 lines.append(output)
         omitted = len(entries) - len(shown)
         if omitted:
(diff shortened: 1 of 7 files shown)


Enough1122:

**Overall:** Correctness fix that stops recovered supervision-loss from being presented as a successful exit. Records `status:"lost"` consistently across poll/log/wait/list and makes gateway concise/coalesced paths warn instead of showing `exit_code:null` as success.

**What it does**
- `tools/process_registry.py:394` in `_refresh_detached_session` sets `session.completion_reason="lost"` + `termination_source="backend_lost"` when re-adopted PID disappears (beyond `exited=True`/`exit_code=None`), adds `_session_status()` (`running`/`lost`/`exited`) and uses it in `poll`/`read_log`/`wait`/`list_sessions`; `read_log`/`poll` now surface `exit_code`/`completion_reason`/`termination_source`.
- `gateway/run.py:3897` updates `_format_concise_process_notification` (`⚠️ supervision lost` vs `✅ finished`/`❌ failed`), `_format_coalesced_process_completions` (header `status updates` vs `completed`, per-entry `supervision lost; no exit...` vs `exit_code=`), and `_run_process_watcher` (always notify on `lost` even under `notify_mode:error`, message wording "supervision was lost; no exit status…" vs "finished with exit code …").
- `hermes_cli/goals.py:1135` narrows `gather_background_processes` / `_render_background_block` to `status=="running"` only (previously `!=exited` included `lost`) — prevents judge seeing lost handles as wait targets.
- `tools/process_registry.py:3214` `format_process_notification` also branches on `reason=="lost"` for coalesced portal path.

**Non-blocking notes**
- `completion_reason` default is still `"exited"` via `getattr(..., "exited")` fallback — existing non-lost sessions remain compatible; ensure any new factory sets field explicitly to avoid default drift.
- Coalesced batch header now varies by presence of any `lost` entry — good, but mixed `lost`+`completed` batches will read as status-update; consider noting mix in header count if UX needs distinction — low priority.
- Change touches only reporting path, not process supervision/killing — no new containment.

fangliquanflq:
Thanks for the review. I checked each note against the current head and no follow-up code change is needed:

- The `completion_reason="exited"` dataclass default is the intentional compatibility value for newly running sessions. The three current `ProcessSession` construction paths start in that ordinary state, while failed starts, kills, backend loss, and recovered-PID disappearance explicitly overwrite the reason before publication (`tools/process_registry.py:391, 1353-1364, 1571-1577, 2435-2438, 2853-2955`).
- `status updates` is deliberately the umbrella header for a mixed batch containing supervision loss. Each shown entry still retains its own outcome (`gateway/run.py:26157-26203`), while all-completion batches retain the existing completion header. The coalesced-delivery tests cover both contracts (`tests/gateway/test_completion_delivery.py:360-396`).
- The reporting-only scope is intentional. Existing checkpoint recovery already re-adopts an identity-verified live host PID as a detached session and preserves status/kill support (`tools/process_registry.py:2853-2955, 2386-2409`); this PR corrects the later transition when that PID can no longer be verified, without changing containment policy.

No code was changed. The current head is `bf723d476a5661ec9a734f887ffa11ac87679090`, and the required CI check is passing.

Issue #90700, opened 2026-08-20
[Bug]: Public /api/status returns unfiltered platform error_message / gateway_exit_reason diagnostics


Atroci:
### Bug Description

`/api/status` is in `PUBLIC_API_PATHS` (intentionally — NAS liveness probing, SPA pre-login bootstrap). But the payload's platform projection only removes two writer-identity keys; every other field inside each `gateway_platforms[*]` entry passes through unfiltered, including `error_message` strings produced by adapter exceptions, plus top-level `gateway_exit_reason`. Adapter error text routinely embeds endpoint URLs, response bodies, and exception details, so whatever an adapter exception included becomes readable by an unauthenticated caller.

### Steps to Reproduce

Against `main` @ fa2601c2f5df58dc33bde47ce1c9cb2f92a75011:

1. `hermes_cli/dashboard_auth/public_paths.py:44` — `/api/status` listed public.
2. `hermes_cli/web_server.py:3649-3653` — entries pass through `_public_platform_entry(value)`.
3. `hermes_cli/web_server.py:3500-3507` — `_public_platform_entry` removes only `_PRIVATE_PLATFORM_ENTRY_KEYS = frozenset({"writer_pid", "writer_start_time"})`. All nested keys (`error_message`, `detail`, arbitrary adapter-specific fields) survive.
4. `hermes_cli/web_server.py:3654, 3797` — `gateway_exit_reason` (a free-text adapter/gateway failure string) is included in the public payload verbatim.
5. `web_server.py:3671-3676` — on `startup_failed`, per-platform `fatal` entries are deliberately *kept* for diagnosis; those are exactly the entries most likely to carry auth-failure/credential-collision detail.

`curl http://<host>/api/status` with no credentials returns `gateway_platforms` with full `error_message` content whenever a platform adapter has failed.

### Expected Behavior

Public projection should allowlist enum/count/state fields per platform entry (the same discipline the `components` rollup already documents at web_server.py:3820-3824: "counts and status enums only"), and `error_message`/`gateway_exit_reason` should be reduced to stable enum-ish codes or omitted on the public path.

### Actual Behavior

Arbitrary nested platform fields and free-text error strings are returned to unauthenticated callers. Impact is conditional (an adapter exception must contain something reusable — URL with embedded credential, token fragment in a response-body echo), but the endpoint is the project's most-exposed surface and the failure mode is silent: nothing forces new adapter fields through any review.

### Affected Component

[...]

PR #97303, opened 2026-08-28
feat(dashboard): withhold deployment detail from anonymous /api/status callers

### LINKED ISSUE #97302: [Feature]: withhold deployment detail (profiles, host memory/disk) from anonymous /api/status callers
### Problem or Use Case

`/api/status` is in `PUBLIC_API_PATHS`, so on a gated (non-loopback) bind any caller who reaches the host reads it without a cookie. That is the documented contract (portal liveness probe, SPA pre-login bootstrap), and the allowlist states the intent: "version, gateway state, active session count, and the dashboard auth-gate shape. No bodies, no session content, no secrets."

The payload today also carries deployment detail:

- `profiles` and `gateway_mode` (`web_server.py`, profile topology), deliberately public because Hermes Cloud renders the profile list in the Portal from an unauthenticated read.
- `memory` (`gateway_rss_mb`, `system_total_mb`, `system_available_mb`, `swap_used_mb`, boot flags) and `disk` (`total_mb`, `free_mb`, `used_percent`).

For a Cloud instance that trade is deliberate. For a **self-hosted** dashboard published on the internet there is no portal reading it, and the same payload publishes the operator's profile roster and host sizing to anyone who curls the hostname. Reproduce with any internet-facing gated dashboard:

```
curl -s https://<host>/api/status | jq '.profiles, .memory, .disk'
```

Related but distinct: #90700 covers free-text `error_message` / `gateway_exit_reason` on the same endpoint. This issue is about the structured deployment fields, which that fix leaves untouched.

### Proposed Solution

A config key, default preserving today's behavior:

```yaml
dashboard:
  public_status_detail: full   # or "minimal"
```

`minimal` withholds `profiles`, `gateway_mode`, `memory` and `disk` from **anonymous** callers on a gated bind. Version, gateway state, `install_id`, session counts and the auth-gate shape stay public, so liveness probes and the desktop's pre-login discovery are unaffected.

For the knob not to blind the operator's own dashboard, the gate's public-path branch should attach `request.state.session` when the caller carries a valid one (best effort, never failing the request), so a signed-in caller keeps the full payload and the Status page keeps rendering its resource banners.

### Alternatives Considered

- Prune unconditionally: regresses the Cloud portal's profile list.
- Gate the fields on `is_managed()`: that flag means package-manager-managed (systemd / Nix), not Cloud-hosted, so it does not identify the reader that needs the fields.
- Front the dashboard with a reverse proxy that filters the response body: works, but every self-hosted operator has to rebuild it, and the endpoint is reachable before any of them think about it.

woprrr:
## What does this PR do?

`/api/status` bypasses the auth gate by design (portal liveness probe, SPA pre-login bootstrap), so on a gated bind anyone who reaches the host reads it cold. Beyond liveness the payload carries deployment detail: `profiles`, `gateway_mode`, and the host `memory` / `disk` rollups. That is intentional for Hermes Cloud, where the Portal renders the profile list from an unauthenticated read. A self-hosted dashboard published on the internet has no such reader, and publishes its profile roster and host sizing to anyone who curls the hostname.

This adds `dashboard.public_status_detail` (default `"full"`, so nothing changes unless an operator opts in). With `"minimal"`, those four fields are withheld from **anonymous** callers on a gated bind; version, gateway state, `install_id`, session counts and the auth-gate shape stay public, so probes and the desktop's pre-login discovery keep working.

For the knob not to blind the operator's own dashboard, the gate's public-path branch now attaches `request.state.session` when the caller carries a valid one, best effort and never failing the request: a public path that 503s because a cookie went stale would be worse than the disclosure it hardens against. A signed-in caller therefore keeps the full payload, and the Status page keeps rendering its resource banners.

Side benefit: on the anonymous path the handler no longer performs the `statvfs` and heartbeat reads at all.

## Related Issue

Fixes #97302

Same endpoint, different field class from #90700 (free-text `error_message` / `gateway_exit_reason`); the two changes do not overlap.

## Type of Change

- [x] 🔒 Security fix

## Changes Made

- `hermes_cli/web_server.py`: `_public_status_detail()` reads `dashboard.public_status_detail` with `HERMES_DASHBOARD_PUBLIC_STATUS_DETAIL` as the env override (same precedence as the other dashboard keys); `get_status` takes the `Request` and computes `withhold_operator_detail` from gate state, the knob, and whether the caller has a session or a token principal. The `request` parameter defaults to `None` so the existing in-process callers that await the handler directly keep working.
- `hermes_cli/dashboard_auth/middleware.py`: `_attach_public_path_session()`, called from the public-path branch. Tries the `Authorization: Bearer` token first (the desktop's native flow holds no cookie, and the public-path branch returns before the gate's own bearer verification), then the session cookie, through the ordered provider stack, swallowing every failure.
- `hermes_cli/config_defaults.py`: the new key, defaulting to `"full"`.
- `cli-config.yaml.example` and `website/docs/user-guide/features/web-dashboard.md`: documented.
- `tests/hermes_cli/test_dashboard_auth_status_endpoint.py`: 9 tests.

## How to Test

Reproduce the exposure, then verify the knob:

1. Run a gated dashboard: `hermes dashboard --host 0.0.0.0 --port 9119 --no-open` with an auth provider configured.
2. `curl -s http://<host>:9119/api/status | jq '.profiles, .memory, .disk'` returns the deployment detail with no credentials.
3. Set `dashboard.public_status_detail: minimal` in `config.yaml`, restart, repeat step 2: the three fields (and `gateway_mode`) are gone, while `version`, `gateway_state`, `auth_required` and `auth_providers` remain.
4. Sign in to the dashboard in a browser, then load `/api/status` in the same session: the full payload is back.

Automated coverage in `tests/hermes_cli/test_dashboard_auth_status_endpoint.py` (9 tests):

- `test_status_publishes_operator_detail_by_default` guards the Cloud default.
- `test_status_withholds_operator_detail_when_minimal` is the leak test.
- `test_status_keeps_operator_detail_for_signed_in_caller` proves the operator is not blinded.
- `test_status_keeps_operator_detail_for_bearer_caller` covers the desktop's cookieless native flow.
- `test_status_minimal_is_a_no_op_on_a_loopback_bind` pins the loopback carve-out.
- `test_status_detail_env_override_wins` and `test_status_detail_unknown_value_falls_back_to_full` cover the setting's precedence and its parsing.
- `test_status_probe_survives_a_broken_session_cookie` and `test_status_survives_an_unreadable_config` pin the never-fail contract of both new code paths.

Each behavior was verified as a genuine regression by reverting it independently: without the middleware cookie attach the signed-in test fails, without the bearer branch the desktop test fails, without the handler change the leak test fails.

### Code

- [x] I've run the test suite for the affected area with `scripts/run_tests.sh`: 50 dashboard / web-server test files, 605 passed. The 24 failures in that set are pre-existing on `main` at the same commit (596 passed / 24 failed before this branch).
- [x] I've tested on my platform: macOS 15 (Darwin 25.6), Python 3.11

- [x] I've updated relevant documentation (`website/docs/user-guide/features/web-dashboard.md`)
- [x] I've updated `cli-config.yaml.example`
- [x] N/A for `CONTRIBUTING.md` / `AGENTS.md`
- [x] Cross-platform: pure Python, no platform-specific primitives touched

### DIFF
--- a/cli-config.yaml.example
+++ b/cli-config.yaml.example
@@ -1888,6 +1888,18 @@ updates:
 #   #     - "172.20.0.5"
 #   #     # - "172.20.0.0/24"  # dedicated network, if the IP is dynamic
 #
+#   # How much of /api/status an ANONYMOUS caller sees on a gated bind.
+#   # /api/status bypasses the auth gate (liveness probe + pre-login
+#   # bootstrap), so on a public host anyone reading the hostname gets this
+#   # payload. "full" (default) keeps every field public, which is what
+#   # Hermes Cloud needs: the Portal renders the profile list from an
+#   # unauthenticated read. "minimal" withholds the deployment detail
+#   # (profiles, gateway_mode, memory and disk rollups) from anonymous
+#   # callers, while a signed-in operator still gets the full payload.
+#   # Override: HERMES_DASHBOARD_PUBLIC_STATUS_DETAIL
+#   #
+#   #   public_status_detail: "minimal"
+#
 # -----------------------------------------------------------------------------
 # Self-hosted OIDC dashboard auth (generic OpenID Connect — Authentik,
 # Keycloak, Zitadel, Authelia, Auth0, Okta, Google, …). Use this INSTEAD of the
--- a/hermes_cli/config_defaults.py
+++ b/hermes_cli/config_defaults.py
@@ -1637,6 +1637,15 @@ DEFAULT_CONFIG = {
         # Set this to True to re-enable the surfaces with the understanding
         # that the numbers are a local lower-bound estimate, not billing.
         "show_token_analytics": False,
+        # How much of ``/api/status`` an ANONYMOUS caller sees on a gated
+        # (non-loopback) bind. "full" keeps every field public, which is what
+        # Hermes Cloud needs: the Portal renders the profile list from an
+        # unauthenticated read. "minimal" withholds the deployment detail
+        # (profile names, gateway mode, host memory / disk rollups) from
+        # anonymous callers while a signed-in operator still gets the full
+        # payload. Meant for a self-hosted dashboard published on the internet,
+        # where nothing but the operator has a reason to read it.
+        "public_status_detail": "full",
         # IP addresses or bounded CIDR networks of reverse proxies allowed to
         # supply X-Forwarded-Proto / X-Forwarded-For. Loopback remains trusted
         # automatically. Wildcards and /0 networks are rejected so arbitrary
--- a/hermes_cli/dashboard_auth/middleware.py
+++ b/hermes_cli/dashboard_auth/middleware.py
@@ -335,7 +335,20 @@ def _attach_public_path_session(request: Request) -> None:
     """
     if getattr(request.state, "session", None) is not None:
         return
+    # The desktop authenticates REST with ``Authorization: Bearer`` and holds
+    # no cookies at all (RFC 8252 native flow), so the bearer has to be read
+    # here too: otherwise a signed-in desktop reads as anonymous on every
+    # public path.
     try:
+        bearer = _extract_bearer(request)
+        if bearer:
+            try:
+                bearer_session = _verify_bearer(request, access_token=bearer)
+            except Exception:
+                bearer_session = None
+            if bearer_session is not None:
+                request.state.session = bearer_session
+                return
         access_token, _refresh_token = read_session_cookies(request)
         if not access_token:
             return
--- a/hermes_cli/dashboard_auth/middleware.py
+++ b/hermes_cli/dashboard_auth/middleware.py
@@ -320,6 +320,38 @@ def _verify_bearer(request: Request, *, access_token: str):
     return None
 
 
+def _attach_public_path_session(request: Request) -> None:
+    """Best-effort session attach for a request on a public (ungated) path.
+
+    Public paths bypass the gate by design, so a handler there cannot tell a
+    signed-in operator from an anonymous probe: both arrive with
+    ``request.state.session`` unset. ``/api/status`` needs that distinction to
+    decide how much deployment detail its payload may carry.
+
+    Nothing here may fail the request: an absent, stale, or unverifiable
+    cookie, or an unreachable IDP, all mean the same thing to the caller of
+    this helper (anonymous), and a liveness probe that 503s because a cookie
+    went stale is worse than the disclosure it was hardening against.
+    """
+    if getattr(request.state, "session", None) is not None:
+        return
+    try:
+        access_token, _refresh_token = read_session_cookies(request)
+        if not access_token:
+            return
+        provider_hint = read_session_provider(request)
+        for provider in _ordered_session_providers(provider_hint):
+            try:
+                session = provider.verify_session(access_token=access_token)
+            except Exception:
+                continue
+            if session is not None:
+                request.state.session = session
+                return
+    except Exception:
+        return
+
+
 async def gated_auth_middleware(
     request: Request,
     call_next: Callable[[Request], Awaitable[Response]],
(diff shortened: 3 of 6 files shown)


Enough1122:

**Overall:** Sound security hardening for self-hosted dashboard on the internet. Keeps default `full` to preserve Hermes Cloud Portal behavior (unauthenticated profile list), while `minimal` withholds deployment detail (`profiles`/`gateway_mode`/`memory`/`disk`) from anonymous callers on gated bind and restores it for valid sessions (cookie or `Bearer`).

**What it does**
- `hermes_cli/config_defaults.py:1637` and `cli-config.yaml.example:1888` add `dashboard.public_status_detail` (`full` vs `minimal`, env `HERMES_DASHBOARD_PUBLIC_STATUS_DETAIL`, env wins, case-insensitive `minimal` check, fail-open to `full` on unreadable config with warning).
- `hermes_cli/dashboard_auth/middleware.py:320` adds `_attach_public_path_session(request)` — best-effort attach of `request.state.session` for public paths (`_path_is_public` branch) via `Bearer` then `Provider.verify_session`, never failing request (stale cookie/bearer/unreachable IDP → anonymous).
- `hermes_cli/web_server.py:3740` adds `_public_status_detail()` and makes `get_status(request, profile)` withhold: `withhold = auth_required and detail==minimal and session is None and not token_authenticated`; when withheld skips `collect_memory_status`/`collect_disk_status` (side benefit: no `statvfs`/heartbeat reads) and omits `status["profiles"]`/`status["gateway_mode"]`; liveness (`version`, `gateway_state`, `install_id`, session counts, auth-gate shape) stays public.
- Docs `website/docs/user-guide/features/web-dashboard.md:398` and tests `tests/hermes_cli/test_dashboard_auth_status_endpoint.py:95` cover default full, minimal anonymous withhold, signed-in keep, broken cookie probe still 200, bearer keeps, loopback no-op, env override wins.

**Non-blocking notes**
- `fail open` on unreadable config is correct for liveness (probe must not 500) but silently drops hardening — logged as warning; ensure log is visible to operator reviewing config.
- `request` param default `None` for in-process callers — treated as anonymous (good seam); ensure no direct internal caller expects full detail and now loses it when unauthenticated.
- `token_authenticated` flag check alongside `session` handles JS SDK case where bearer verified but session shape differs — complete.

woprrr:
Thanks for the review. Verified each note against the branch; one was actionable and is now fixed.

**Fail-open warning visibility.** You were right to look at it, and it was worse than "is it visible": the warning sat on a per-request path. `/api/status` is polled on a timer by portals, uptime monitors and the desktop, so a single unreadable config would emit one warning per probe and bury the line the operator needs. Now warned once per process, following the same one-shot discipline as the malformed-`public_url` warnings in `dashboard_auth/prefix.py`. A test asserts three probes produce exactly one warning, and it fails without the guard. Pushed.

**In-process callers losing detail.** Checked: the only direct caller of `web_server.get_status` outside the ASGI route is `tests/hermes_cli/test_web_server.py`, which asserts the config load stays off the event loop and does not read `profiles` / `memory` / `disk`. Every other consumer arrives over HTTP and therefore carries a `Request`. No production caller loses anything.

**`token_authenticated`.** Small correction: that flag is set only for routes registered via `register_token_route` (today the drain endpoint), and `/api/status` is not one of them, so it never fires there. The bearer case is handled by `_attach_public_path_session`, which verifies the `Authorization: Bearer` token through the same provider stack before falling back to the cookie. The flag stays in the condition as defence in depth if `/api/status` ever becomes token-authable, and `test_status_keeps_operator_detail_for_bearer_caller` covers the real path.

One nit for future automated passes: the line anchors in the summary have drifted. The new key is `config_defaults.py:1648`, the helper is `middleware.py:323`, `_public_status_detail` is `web_server.py:3743`, and the example config entry is `cli-config.yaml.example:1901`.

PR #97304, opened 2026-08-28
fix(agent): stop incomplete runs from reporting success

### LINKED ISSUE #97278: [Bug]: run_conversation exits bypass finalize_turn — no Turn ended line, and the run still surfaces as completed with empty output
### Summary

`run_conversation` has **26 `return` statements that bypass `finalize_turn`** and exactly **one** that goes through it. Because `finalize_turn` is what emits the `Turn ended` log line, a turn that leaves through any of the 26 produces no terminal log record at all — and the caller cannot tell which of the 26 fired, or that one fired at all.

Every one of those returns carries `"completed": False` (most also `"partial": True` and an `"error"` string). We observe the corresponding run surfacing through the Runs API as **terminal with empty output**, with its last recorded event still a tool event rather than a run-level one. A consumer polling that API cannot distinguish *"the turn finished and had nothing to say"* from *"the turn bailed out on one of 26 error paths"*.

This is an observability report, not a claim that any single path is wrong. Two of the 26 already carry comments acknowledging the bypass is deliberate:

```python
# Prior successful tool batches (or injected tool
# errors) can leave a tool-result tail; this path
# never reaches finalize_turn (#48879 class).
```
— `agent/conversation_loop.py:4071` (HEAD)

```python
# Same tool-tail close as interrupt / invalid-tool
# exhaustion — this path never reaches finalize_turn.
```
— `agent/conversation_loop.py:7183` (HEAD)

The request is that leaving by these paths be **visible**, not that they be removed.

### Environment

- **Observed on:** Hermes Agent v0.19.1, `HERMES_AGENT_REF=v2026.7.30`, commit `cc4cab2f592e60a197e796506de9168f74baf3ea`
- **Verified structurally against HEAD** (`eff97a8a05`, after v2026.8.27): 24 bypasses at our pin, **26 at HEAD** — the shape has grown, not gone
- `api_mode: chat_completions` against a self-hosted OpenAI-compatible endpoint
- Consumer is a long-running harness polling `GET /v1/runs/{id}`

We have **not** re-run our workload on v2026.8.27, so this is a structural verification at HEAD plus behavioural evidence at our pin. Flagging that explicitly rather than implying we reproduced on current.

### Evidence

Across four log rotations covering 2026-08-27 22:49Z to 2026-08-28 15:31Z, **5 of 29 sessions have no `Turn ended` line at all**.

The most informative of the five delivered **1,540 characters of output** and still logged no terminal line. That is what points at the bypass returns rather than at some quiet-model failure mode: a return whose `final_response` is `partial_response or None` yields content whenever the partial is non-empty, and logs nothing either way.

Sessions torn down by a cancellation are not among the five, and not because they were filtered out — an interrupted session still reaches `finalize_turn` and logs `Turn ended: reason=interrupted_by_user` or `interrupted_during_api_call`. Both interrupted sessions in the corpus did. The five returned without reaching finalisation at all.

Tracing one of the five, the one that does trail off, to the point it stops:

```
04:50:54,344  API call #15: model=<route> out=130 latency=4.8s
04:50:54,387  tool skill_view completed (2470 chars)
04:50:54,392  WARNING Pre-call sanitizer: healed 3 empty non-final message(s)
              — nothing further. API call #16 never happens.
04:52:50,939  gateway GET /v1/runs/run_8560c… → 200, 832 bytes (prior polls: 732)
              — the run leaves the poll set.
```

No error, no traceback, no `OpenAI client closed`. A session that finalises normally ends with `Turn ended` → `OpenAI client closed (shared=True)` → `(shared=False)`.

That session trails off in output before leaving — per-call output tokens `1280, 239, 343, 705, 425, 629, 596, 160, 163, 151, 294, 130`, latency 35.4s → 4.8s. **Three of the five do the opposite**, ending on 6,307, 4,764 and 9,613 output tokens. So the exit does not correlate with the model running down, and we are not reporting a quiet-model failure mode. Included because the trailing-off case may be a distinct sub-mode worth separating.

From our own job store — a wider window than the log above, which has rotated — **nine unit-runs across four separate jobs** produced no output at all. **Two of those four jobs were recorded successful anyway**, because empty output from a terminal run was indistinguishable from a legitimately quiet one. On one job the downstream stage then spent 164 minutes rediscovering from the filesystem what the silent unit had never reported.

The two counts are different windows on the same failure and not the same set: 5 of 29 is sessions-without-a-terminal-line in the surviving log; nine unit-runs is units-with-no-output in the store. One session had no terminal line and did produce output, so neither count contains the other.

### What we ruled out

Recording these because each looked right and cost a round of investigation:

- **Turn ended on a tool call.** No `Turn ended` line in the corpus ends on a tool call: of the 24 that exist, 22 read `reason=text_response(finish_reason=stop)` and 2 read `interrupted_by_user` / `interrupted_during_api_call` from an operator cancellation.
- **Context exhaustion.** The failing session died at `in=44,729` against a declared input budget of 64,512 and a served window of 86,016 — 69% of budget, 52% of window. A healthy session on a larger lane ran to `in=96,258` and finished.
- **A wall-clock timeout.** The nine runs span 786–3,319 seconds against a 14,400-second cap; the longest reached 23% of it.
- **`max_iterations`.** Our sessions run against a ceiling of 500 — `api_calls=N/500` throughout — and the five used 5 to 15 calls. The `16` that also appears in our lines belongs to `background_review.py`, a separate thread.

### One thing we could not rule out

**Empty-content assistant turns may be involved, and our first reading of this was backwards.** `Pre-call sanitizer: healed N empty non-final message(s)` fires 75 times across the 22 sessions that finalise and 41 times across the 5 that do not, which looks like it fires more often where things work — until you divide by session count. Per session it is **3.4 against 8.2**, more than twice as often on the runs that never finalise.

We cannot tell from outside whether that is cause, effect, or a shared upstream of both, and 5 sessions is a thin base. Reporting it as an open lead rather than an exclusion because the raw totals had it in the "ruled out" column and the normalised figures do not support that.

### Relationship to existing issues

Checked before filing; this is adjacent to three open issues but distinct from each:

- **#83915** — length-truncation continuation, ending at the `partial_response or None` return (`conversation_loop.py:4004` at HEAD, message `Response remained truncated after 4 continuation attempts`). That is **one of the 26**, and notably the one that can return `final_response: None`. Our profile does not match it: #83915 requires four full-budget generations, and our failing calls run 130–294 output tokens. Same family, different member — which is the point of this issue.
- **#84207** — interrupted turn (exit 130) delivers zero user feedback. Close cousin, but that path **does** emit `Turn ended: reason=interrupted_by_user … response_len=0`. Ours emits nothing, so it is a different exit.
- **#48879** (closed) — cited in the code comments above for the tool-result-tail close, not for the finalisation bypass itself.

### Suggested direction

Any of these would resolve the operator-visible half:

1. **Emit a terminal log line on the bypass paths** — the same shape as `Turn ended`, carrying which path fired and the `error` string already in the return dict. Cheapest fix; makes all 26 greppable.
2. **Route the bypasses through `finalize_turn`** with a reason argument, so there is one exit that logs. Larger change, and #78641 / #49864 suggest that area is already being reconsidered.
3. **Propagate `completed: False` / `partial: True` to the Runs API** so a polling consumer sees a non-clean terminal state rather than a completion with empty output. This is the half that made us mis-record work as successful.

(1) and (3) are independent and (3) is what actually cost us; a run that reports terminal-successful with no output is the failure mode that ships.

### What we did downstream, for context

Our worker now treats a terminal run whose last event is not run-level as unverified rather than successful, and refuses it. That converts a silent wrong answer into a loud failure our retry path can act on. It does not make the unit succeed, and it is a workaround for the missing signal rather than a fix.

fangliquanflq:
## What does this PR do?

Incomplete conversation exits now leave a terminal diagnostic and propagate as failed `/v1/runs` records instead of terminal-successful runs with empty or partial output. The fix classifies the existing `completed`, `failed`, and `partial` result contract at the two outer boundaries, avoiding behavior changes inside the 26 early-return paths.

### Symptom

An early `run_conversation` return can bypass `finalize_turn`, so no `Turn ended` line is logged. `/v1/runs` then checks only `failed=True`; results carrying `completed=False` or `partial=True` are published as `run.completed`, including empty output.

### Impact

Polling consumers cannot distinguish a clean quiet completion from an aborted turn and can record incomplete work as successful. Partial text can also be presented as a complete run.

### Bug Cause

**Trigger:** `run_agent.py:8966` / the outer `AIAgent.run_conversation` boundary, followed by `gateway/platforms/api_server.py:7886` / `_handle_runs`.

**Causal chain:**
1. A conversation-loop recovery or exhaustion path returns a structured result with `completed=False`, often with `partial=True`, without reaching `finalize_turn`.
2. The outer wrapper treats any non-interrupted result lacking `failed=True` as a successful relay outcome and emits no replacement terminal diagnostic.
3. `/v1/runs` also checks only `failed=True`, publishes `run.completed`, and stores a terminal `completed` status.

**Why it is wrong:** `completed=False` and `partial=True` are explicit non-success signals in the result contract, but both consumers ignore them.

**Working sibling / contrast:** `/v1/chat/completions` already evaluates `completed`, `failed`, and `partial` together and returns an incomplete/error response instead of silently succeeding.

**Ruled out:** This is not iteration-budget, wall-clock, or cancellation handling. The failure reproduces deterministically by returning an incomplete result dict directly from the agent boundary.

### Fix

- Classify `failed=True`, `completed=False`, or `partial=True` as incomplete at the outer conversation boundary.
- Stamp bypass results with `turn_exit_reason=early_return`, emit one redacted `Turn ended` warning, and close relay instrumentation with a failed outcome.
- Publish `/v1/runs` incomplete results as `run.failed`, preserving redacted error text, partial output, and machine-readable completion flags.
- Add behavior tests for both empty and non-empty partial failures.

## Related Issue

Closes #97278

## Type of Change

## Changes Made

- `run_agent.py` - diagnose and classify early incomplete returns at the shared outer boundary.
- `gateway/platforms/api_server.py` - map every explicit incomplete result to failed run status and events.
- `tests/run_agent/test_run_agent.py` - verify the terminal log and failed relay outcome.
- `tests/gateway/test_api_server_runs.py` - verify failed status for empty and partial-output results.

## How to Test

1. Return `completed=False` from a mocked conversation and confirm the wrapper logs `Turn ended` and reports a failed relay outcome.
2. Start `/v1/runs` with empty and non-empty incomplete results and poll until the status is terminal.
3. Run:

```bash
scripts/run_tests.sh tests/run_agent/test_run_agent.py -k early_incomplete_return_logs_terminal_failure
scripts/run_tests.sh tests/gateway/test_api_server_runs.py
```

The focused wrapper test passed (1 test), and the full `/v1/runs` file passed (27 tests). A broader `tests/run_agent/test_run_agent.py` run passed 279 tests; its unrelated Anthropic interrupt test could not construct the optional provider because the local environment does not have the `anthropic` package installed.

### Code

- [x] I've run the relevant repository test entry points and documented the results above

- [x] Documentation update: N/A; the behavior contract is covered by code comments and tests
- [x] Config example update: N/A; no config keys changed
- [x] Architecture/workflow docs update: N/A; no architecture or workflow changed
- [x] Cross-platform impact considered: the change is platform-independent result classification
- [x] Tool descriptions/schemas update: N/A; no tool behavior changed

## Screenshots / Logs

N/A; automated status and logging assertions cover the regression.

### DIFF
--- a/gateway/platforms/api_server.py
+++ b/gateway/platforms/api_server.py
@@ -7880,21 +7880,35 @@ class APIServerAdapter(BasePlatformAdapter):
                         "cancelled",
                         last_event="run.cancelled",
                     )
-                # Check for structured failure (non-retryable client errors like
-                # 401/400 return failed=True instead of raising, so the except
-                # block below never fires — issue #15561).
-                elif isinstance(result, dict) and result.get("failed"):
+                # Structured non-success results return normally instead of
+                # raising. Treat every explicit incomplete contract as failed;
+                # checking only ``failed`` turned completed=False / partial=True
+                # early exits into terminal-successful empty runs (#97278).
+                elif isinstance(result, dict) and (
+                    result.get("failed") is True
+                    or result.get("completed") is False
+                    or result.get("partial") is True
+                ):
                     error_msg = _redact_api_error_text(result.get("error") or "agent run failed")
+                    final_response = result.get("final_response", "") or ""
+                    completed = result.get("completed") is True
+                    partial = result.get("partial") is True
                     _put_event_if_active({
                         "event": "run.failed",
                         "run_id": run_id,
                         "timestamp": time.time(),
                         "error": error_msg,
+                        "output": final_response,
+                        "completed": completed,
+                        "partial": partial,
                     })
                     self._set_run_status(
                         run_id,
                         "failed",
                         error=error_msg,
+                        output=final_response,
+                        completed=completed,
+                        partial=partial,
                         last_event="run.failed",
                     )
                 else:
--- a/run_agent.py
+++ b/run_agent.py
@@ -8985,9 +8985,37 @@ class AIAgent:
                     # outer finally: a refresher firing between stop and join
                     # would otherwise set an interrupt that survives the clear.
             terminal = result if isinstance(result, dict) else {}
+            terminal_incomplete = (
+                terminal.get("failed") is True
+                or terminal.get("completed") is False
+                or terminal.get("partial") is True
+            )
+            if terminal_incomplete and not terminal.get("turn_exit_reason"):
+                # Early returns inside conversation_loop intentionally bypass
+                # finalize_turn, but that must not also erase the terminal
+                # diagnostic operators use to distinguish a clean empty answer
+                # from an aborted turn (#97278).  Stamp the shared result at the
+                # outer boundary so every present and future bypass is covered.
+                terminal["turn_exit_reason"] = "early_return"
+                _early_error = redact_sensitive_text(
+                    str(terminal.get("error") or terminal.get("failure_reason") or "")
+                )[:500]
+                logger.warning(
+                    "Turn ended: reason=early_return completed=%s failed=%s "
+                    "partial=%s interrupted=%s api_calls=%s response_len=%d "
+                    "session=%s error=%s",
+                    terminal.get("completed"),
+                    terminal.get("failed") is True,
+                    terminal.get("partial") is True,
+                    terminal.get("interrupted") is True,
+                    terminal.get("api_calls", "unknown"),
+                    len(str(terminal.get("final_response") or "")),
+                    session_id or "none",
+                    _early_error or "none",
+                )
             if terminal.get("interrupted") is True:
                 relay_outcome = "cancelled"
-            elif terminal.get("failed") is True:
+            elif terminal_incomplete:
                 relay_outcome = "failed"
             else:
                 relay_outcome = "success"
--- a/tests/gateway/test_api_server_runs.py
+++ b/tests/gateway/test_api_server_runs.py
@@ -154,6 +154,46 @@ class TestStartRun:
                 assert status["object"] == "hermes.run"
 
     @pytest.mark.asyncio
+    @pytest.mark.parametrize(
+        ("final_response", "partial"),
+        [("", False), ("partial answer", True)],
+    )
+    async def test_incomplete_agent_result_reports_failed_run(
+        self, adapter, final_response, partial
+    ):
+        app = _create_runs_app(adapter)
+        async with TestClient(TestServer(app)) as cli:
+            with patch.object(adapter, "_create_agent") as mock_create:
+                mock_agent = MagicMock()
+                mock_agent.run_conversation.return_value = {
+                    "final_response": final_response,
+                    "completed": False,
+                    "partial": partial,
+                    "error": "response truncated",
+                }
+                mock_agent.session_prompt_tokens = 10
+                mock_agent.session_completion_tokens = 5
+                mock_agent.session_total_tokens = 15
+                mock_create.return_value = mock_agent
+
+                response = await cli.post("/v1/runs", json={"input": "hello"})
+                run_id = (await response.json())["run_id"]
+
+                for _ in range(40):
+                    status_response = await cli.get(f"/v1/runs/{run_id}")
+                    status = await status_response.json()
+                    if status["status"] == "failed":
+                        break
+                    await asyncio.sleep(0.05)
+
+        assert status["status"] == "failed"
+        assert status["last_event"] == "run.failed"
+        assert status["completed"] is False
+        assert status["partial"] is partial
+        assert status["output"] == final_response
+        assert status["error"] == "response truncated"
+
+    @pytest.mark.asyncio
     async def test_start_binds_chat_id_for_delegation_wake_target(self, adapter):
         """/v1/runs must bind the raw session id as the api_server chat_id
         (like every other agent-entry route does via _run_agent): the async
--- a/tests/run_agent/test_run_agent.py
+++ b/tests/run_agent/test_run_agent.py
@@ -3267,6 +3267,45 @@ class TestRunConversation:
         assert result is failed_result
         assert order == ["logical", "metrics"]
 
+    def test_early_incomplete_return_logs_terminal_failure(self, agent):
+        from agent import relay_runtime
+
+        incomplete_result = {
+            "final_response": "partial answer",
+            "messages": [],
+            "api_calls": 2,
+            "completed": False,
+            "partial": True,
+            "error": "response truncated",
+        }
+
+        with (
+            patch(
+                "agent.conversation_loop.run_conversation",
+                return_value=incomplete_result,
+            ),
+            patch(
+                "hermes_cli.observability.relay_shared_metrics.start_task_run",
+            ),
+            patch(
+                "hermes_cli.observability.relay_shared_metrics.finish_task_run",
+            ),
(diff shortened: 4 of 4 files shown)


Enough1122:

**Overall:** Important contract fix at two outer boundaries that stops empty/partial incomplete turns from being recorded as `run.completed`. Classifies existing `completed/failed/partial` signals without touching the 26 inner early-return sites.

**What it does**
- `run_agent.py:8980` at outer `run_conversation` finally computes `terminal_incomplete = failed is True or completed is False or partial is True`; if incomplete and `turn_exit_reason` absent, stamps `terminal["turn_exit_reason"]="early_return"` and `logger.warning("Turn ended: reason=early_return completed=…")` (redacted error ≤500), and maps `relay_outcome` to `failed` (vs cancelled/success) — ensures `finish_logical_calls(outcome=failed)` and metrics reflect failure even when `finalize_turn` was bypassed.
- `gateway/platforms/api_server.py:7880` in `_run_sync` widens `elif failed:` to `(failed is True or completed is False or partial is True)`, propagates `final_response` as `output`, `completed`/`partial` into both `run.failed` event and `run status=failed` (mirroring `/v1/chat/completions` already does).
- Tests `tests/gateway/test_api_server_runs.py:126` param `("",False)` and `("partial answer",True)` assert `status==failed`, `completed is False`, etc.; `tests/run_agent/test_run_agent.py:3267` asserts `finish_logical_calls(outcome==failed)` and `Turn ended:` warning.

**Non-blocking notes**
- Strict `is True` / `is False` checks avoid truthy coercion — e.g. `completed=None` not treated as incomplete (only explicit `False`) — matches intended contract; ensure older callers never emit `"false"` string.
- Stamping `turn_exit_reason` only when missing avoids overwriting genuine reason from inner path; good scoping.
- Login to `logger.warning` includes `response_len` and `api_calls` — helpful for empty-output diagnosis.

fangliquanflq:
Thanks for the review. I checked each note against the current head:

- The strict identity checks are intentional. All `run_conversation` result producers in the affected paths emit real booleans; a repository search found no producer using string `"true"`/`"false"` values. The early-return paths in `agent/conversation_loop.py` consistently use `completed: False` and/or `partial: True`.
- `turn_exit_reason` is only added when absent, so an existing, more specific reason remains authoritative.
- The warning remains metadata-only and redacts/truncates the error while retaining `api_calls` and response length for diagnosis.

No code change is needed. I re-ran the focused regression test and the full Runs API test file with the canonical runner: `scripts/run_tests.sh tests/run_agent/test_run_agent.py -k early_incomplete_return_logs_terminal_failure` (1 passed) and `scripts/run_tests.sh tests/gateway/test_api_server_runs.py` (27 passed). Current required CI is also green.

Issue #97305, opened 2026-08-28
Photon rich-link sends fail with target_not_allowed on never-contacted chats (plan limitation, asymmetric with plain-text sends)


semirkabir:
## Summary

Outbound rich-link sends (`/send-richlink`, and any URL-only message routed through `spectrum-ts` `richlink()`) fail with `AuthenticationError: Target not allowed for this project` when the target chat has never received an outbound message from that Photon project. The adapter already classifies this correctly (`target_not_allowed`, non-retryable — see `classifySidecarError` in the sidecar), but the failure mode is surprising for users because:

1. The same URL sends fine as **plain text** via `/send` (markdown/text path does not hit the richlink builder's space-permission check the same way).
2. The same URL sends fine via **richlink** once the chat is "warmed up" with any successful outbound send.

## Reproduction

On a shared/free-tier Photon project line (relay `@photon-ai/advanced-imessage`):

```
POST /send-richlink  {spaceId: "<never-contacted E.164>", url: "https://x.com/..."}
→ {"ok":false,"error":"internal sidecar error","error_class":"target_not_allowed","retryable":false}

POST /send  {spaceId: "<same id>", text: "control message"}
→ {"ok":true, ...}   (this may itself fail with target_not_allowed on a truly cold target —
                      but on a chat with prior inbound history the plain send succeeds
                      while the richlink send does not)

POST /send-richlink  {spaceId: "<same id>", url: "..."}   # after one successful send
→ {"ok":true, ...}
```

Verified live 2026-08-28 on a deployment running current `main` adapter code: identical URL, identical spaceId — plain-text send succeeds, richlink send throws `Target not allowed for this project` until a prior successful outbound exists.

## Root cause

spectrum-ts raises `AuthenticationError("Target not allowed for this project")` from `space.send()` when a shared/free-tier project tries to **initiate** an outbound conversation with a new target. The `richlink()` builder path enforces this check on every call, while the plain-text path only enforces it on genuinely cold targets. The result is an asymmetry that looks like a bug in Hermes' rich-link routing (`_richlink_candidate` → `_sidecar_send_richlink`) but is actually a relay plan limitation surfaced inconsistently.

The sidecar already does the right thing classifying it (`target_not_allowed`, `retryable: false`, issues #50971/#51897 context) and the adapter logs and falls back to plain text — the user-visible symptom is just that URL-only messages to new chats occasionally drop their preview card for no obvious reason.

## Suggested improvements

[...]

PR #97306, opened 2026-08-28
fix(photon): explain rich-link target_not_allowed fallback and carry error class

semirkabir:
## Summary

Follow-up to #97305. Rich-link sends to never-contacted targets fail with
`Target not allowed for this project` on shared/free-tier Spectrum projects,
and the adapter's generic warning (`rich-link send failed, falling back to
plain text`) made the plan limitation look like a Hermes routing bug. This PR
makes the failure self-explanatory and documented.

## Changes

**`plugins/platforms/photon/adapter.py` — `_sidecar_send_richlink`**
- Distinguishes `PhotonSidecarError.error_class`:
  - `target_not_allowed` → logs an **info**-level line that says what actually
    happened ("target not warmed for rich-link previews — sending as plain
    text; rich-link previews should work on future sends to this chat")
    instead of a scary warning.
  - anything else → keeps the existing warning.
- Propagates `error_class` / `retryable` onto the returned `SendResult`
  (`raw_response` + `retryable`), mirroring what `_sidecar_send` already does
  for `PhotonSidecarError`, so callers can distinguish a plan limitation from
  a transient fault.

**`plugins/platforms/photon/README.md`**
- Documents the warm-up behavior under the rich-link bullet: never-contacted
  chats can reject `richlink()` sends, the message still lands as plain text,
  and the first successful outbound send warms the target up.

**Tests**
- New `test_target_not_allowed_falls_back_and_reports_error_class`: the
  fallback chain still delivers the URL as plain text, and the richlink
  attempt reports `{"error_class": "target_not_allowed", "retryable": False}`.

## Validation

```
pytest tests/plugins/platforms/photon/test_rich_links.py  → 8 passed
pytest tests/plugins/platforms/photon/                    → 129 passed,
  1 pre-existing failure on pristine main
  (test_spectrum_patch.py::test_sidecar_patch_failure_still_reaches_health_endpoint,
  fails identically without this change in this container)
```

Refs #97305

### DIFF
--- a/plugins/platforms/photon/README.md
+++ b/plugins/platforms/photon/README.md
@@ -156,6 +156,14 @@ All env vars are documented in `plugin.yaml`. The most important:
   via spectrum-ts' `richlink()` builder so iMessage can render a native link
   preview card. `PHOTON_MARKDOWN=false` reverts to stripped plain text and
   disables rich-link routing.
+
+  Rich-link sends to a chat that has never received a successful outbound
+  message can fail with `Target not allowed for this project` on
+  shared/free-tier Spectrum projects — spectrum-ts enforces the
+  outbound-initiation check on every `richlink()` call. When this happens the
+  adapter falls back to plain text and logs a warning; the first successful
+  outbound send to a target warms it up and rich-link previews work from then
+  on (see #97305).
 - **Reactions (tapbacks) are supported** behind `PHOTON_REACTIONS` (default
   off): the adapter tapbacks 👀 while processing and swaps it for 👍/👎 on
   completion, and a user tapback on a bot-sent message is routed to the agent
--- a/plugins/platforms/photon/adapter.py
+++ b/plugins/platforms/photon/adapter.py
@@ -2503,6 +2503,33 @@ class PhotonAdapter(BasePlatformAdapter):
             data = await self._sidecar_call(
                 "/send-richlink", {"spaceId": space_id, "url": url}
             )
+        except PhotonSidecarError as e:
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** UX/docs fix that makes a known Spectrum shared-project plan limitation self-explanatory instead of looking like a routing bug. Follow-up to #97305, well documented and minimally invasive.

**What it does**
- `plugins/platforms/photon/adapter.py:2503` in `_sidecar_send_richlink` catches `PhotonSidecarError` and branches on `error_class=="target_not_allowed"` → `logger.info` with warm-up explanation vs generic `logger.warning "rich-link send failed"`; returns `SendResult(success=False, error=str(e), raw_response={"error_class","retryable"}, retryable=e.retryable)` (mirrors `_sidecar_send` pattern).
- `plugins/platforms/photon/README.md:52` documents warm-up behavior under rich-link bullet (never-contacted chats can reject `richlink()`, fallback to plain text, first successful outbound warms target).
- Test `tests/plugins/platforms/photon/test_rich_links.py:106` covers `target_not_allowed` still delivering via `"/send"` fallback and `raw_response`/`retryable==False`.

**Non-blocking notes**
- `info` vs `warning` distinction is appropriate — plan limitation is expected, not operational anomaly. Ensure log aggregation filters treat `target_not_allowed` info as non-alert.
- `raw_response` exposure is limited to `error_class`/`retryable` — no secret leakage; fallback path still succeeds (`success=True` with `message_id`).

PR #97307, opened 2026-08-28
fix(update): refresh Linux desktop entry when build is current

kasimali59:
## Summary
- Refresh the Linux desktop entry even when the packaged desktop build is already current.
- Prevent stale taskbar launchers from surviving Hermes updates.
- Add regression coverage for the skipped-build path.

## Test plan
- `uv run --directory /tmp/hermes-agent-upstream-fix --with pytest python -m pytest -q tests/hermes_cli/test_cmd_update.py::TestNodeRuntimeNpmResolution::test_update_refreshes_linux_desktop_entry_when_build_is_current tests/hermes_cli/test_linux_desktop_entry.py`
- Result: 16 passed, 2 skipped
- `git diff --check`

The two skipped tests are platform/environment-specific.

### DIFF
--- a/hermes_cli/linux_desktop_entry.py
+++ b/hermes_cli/linux_desktop_entry.py
@@ -78,11 +78,11 @@ def resolve_exec_command() -> str:
             # third-party import (#90292) — silently, since Terminal=false.
             # sys.executable is the interpreter actually running Hermes (the
             # venv one), so prefix it explicitly.
-            argv = [str(Path(sys.executable).resolve()), str(resolved), "desktop"]
+            argv = [str(_interpreter_path()), str(resolved), "desktop"]
         else:
             argv = [str(resolved), "desktop"]
     else:
-        argv = [str(Path(sys.executable).resolve()), "-m", "hermes_cli.main", "desktop"]
+        argv = [str(_interpreter_path()), "-m", "hermes_cli.main", "desktop"]
     return " ".join(_quote_exec_arg(a) for a in argv)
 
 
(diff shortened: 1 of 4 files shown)


Enough1122:

**Overall:** Corrective self-healing fix that prevents stale Linux `.desktop` launcher from surviving updates when Electron build is already current. Keeps the "Desktop up to date" fast path but still repairs the entry.

**What it does**
- `hermes_cli/update_cmd.py:7165` in `_rebuild_desktop_after_update` when `skip_desktop_build` true now calls `_m()._register_linux_desktop_entry()` before `return True` (still prints "Desktop app up to date").
- `hermes_cli/linux_desktop_entry.py:38` adds `_interpreter_path()` (returns `Path(sys.executable)` absolute-venv-preserving vs `resolve()`) and uses it in `resolve_exec_command()` and `_needs_interpreter()` (`exe_dir` now `_interpreter_path().parent.lower()`), preserving venv discovery for venv-symlinked interpreter.
- Tests `tests/hermes_cli/test_cmd_update.py:1007` pin `skip_build` still calls register; `tests/hermes_cli/test_linux_desktop_entry.py:107` update exec prefix assertion to `absolute()` vs `resolve()`.

**Non-blocking notes**
- `sys.prefix != sys.base_prefix` correctly detects venv vs system interpreter; ensure `is_absolute()` fallback covers relative `sys.executable` (rare but possible when launched via `PATH` wrapper).
- Call is Linux-only but gated inside function — `_register_linux_desktop_entry` is no-op on non-Linux — safe to invoke unconditionally.

kasimali59:
Addressed the non-blocking review note about relative `sys.executable` values.

- Added an explicit canonical `.resolve()` fallback for relative interpreter paths.
- Added regression coverage for the relative-path case.
- Verification: clean upstream worktree launcher/update tests pass (17 passed, 2 skipped).

kasimali59:
@Enough1122 Thanks for the note. Addressed in the latest update:

- Relative `sys.executable` values now use an explicit canonical `.resolve()` fallback.
- Absolute venv interpreter paths retain their venv identity so Hermes dependencies remain available.
- Added regression coverage for the relative-path case.

Verification passes in both environments: local launcher tests 16 passed, 2 skipped; clean upstream worktree 17 passed, 2 skipped.

Enough1122:
Verified at head `fd7ccd3f77` — addressed exactly as described. `_interpreter_path()` is the right shape: relative `sys.executable` gets the canonical `.resolve()` fallback (the new `test_relative_interpreter_path_uses_resolved_absolute_fallback` pins it), while an absolute venv interpreter retains its venv path (`sys.prefix != sys.base_prefix`) so Python still discovers the venv's packages. The extra self-healing in `_rebuild_desktop_after_update` (re-registering the Linux desktop entry even when the Electron build is current) is a good catch — a stale entry keeping a retired Python wrapper alive after update is exactly the failure mode the launcher should defend against.

No further items from me.

kadiratesdev:
Current main now contains the broader venv-aware Exec resolution changes, leaving this PR conflicted while the update fast-path hook is still needed. I ported only that remaining hook and its regression test onto current main in #99755. The commit preserves @kasimali59 as author, and the new PR credits this work explicitly.

alt-glitch:

Duplicate of #99755. That clean current-main port carries the same remaining update fast-path hook and credits this work; the broader launcher-resolution portion of this PR is already on main.

PR #29013, opened 2026-05-20
feat(skills): add auto_sync_bundled config to disable bundled skill auto-copy

ScotterMonk:
## Summary

Add `skills.auto_sync_bundled` config flag (default: `true`) to control whether new bundled skills added upstream are automatically installed into `~/.hermes/skills/` during sync/update.

When set to `false`:
- Newly-added upstream skills are **not** auto-copied
- Existing already-tracked skills still update normally
- User-deleted skills remain respected
- Explicit `hermes skills reset <name>` / `hermes skills reset <name> --restore` still work (bypass the flag)
- Re-enabling the flag picks up skipped skills on the next sync

## Why

The existing manifest system respects user deletions but cannot prevent newly introduced bundled skills from being auto-installed. Users who intentionally keep a minimal local skills set find unwanted skills (e.g. platform-specific skills like `apple-notes` on Linux) silently appearing after upstream updates. The per-profile `.no-bundled-skills` marker file is all-or-nothing — there is no equivalent for the default profile or for users who want to keep existing bundled skills updated while blocking only new additions.

## Changes

- **`hermes_cli/config.py`**: Add `auto_sync_bundled: true` to skills config defaults
- **`tools/skills_sync.py`**: 
  - New `_load_auto_sync_bundled()` helper reads config.yaml directly (avoids import coupling with hermes_cli.config)
  - New `_sync_one_bundled_skill()` force-syncs a single skill, bypassing auto_sync suppression (for reset/restore)
  - `sync_skills()` returns `skipped_auto` count when auto-sync suppresses new skills
  - `reset_bundled_skill()` uses `_sync_one_bundled_skill` so explicit restores work regardless of the flag
- **`hermes_cli/main.py`**: Show `skipped_auto` count in update output
- **`tests/tools/test_skills_sync.py`**: 7 new test cases covering:
  - auto_sync=false skips new skills
  - auto_sync=true copies new skills (default behavior)
  - String `"false"` parsed correctly
  - Disable-then-reenable picks up skipped skills
  - Missing config defaults to true
  - Missing bundled dir returns `skipped_auto: 0`
  - Reset/restore bypasses auto_sync suppression

Closes #11894

See also: #19986 (broader discussion about which skills should be bundled)

Issue #19986, opened 2026-05-05
Make non-core bundled skills optional and keep the default install minimal


WompaJango:
### Problem or Use Case

Hermes currently ships a large set of bundled skills directly into each profile's skills directory via bundled-skill sync. That makes updates heavier, expands the default install footprint, and creates an ongoing maintenance burden for content that many users will never use.

A concrete example is the Apple/macOS skill set (`apple-notes`, `apple-reminders`, `findmy`, `imessage`). These are valid skills, but on non-macOS systems they still exist as bundled content and can resurface through sync/update behavior even though they are platform-specific and irrelevant for most installs.

This creates a few problems:

- The default install is not minimal; it includes many non-core skills by default.
- Bundled skill sync can reintroduce content users intentionally removed.
- Every bundled SKILL.md becomes part of the long-term maintenance surface and can drift from reality.
- Platform-specific or niche skills feel like product defaults instead of opt-in extensions.

The current model makes sense for a very small core set of universally useful skills, but it scales poorly as the catalog grows.

### Proposed Solution

I think Hermes should distinguish between **core bundled skills** and **optional official skills** much more aggressively.

Suggested approach:

1. Define a small core set of bundled-by-default skills
   - Only keep skills that are broadly useful across platforms/workflows.
   - Everything else should move to `optional-skills/` or another official opt-in channel.

2. Make non-core bundled skills install-on-demand
   - `hermes skills install <official-skill>` should be the normal path for niche/platform-specific skills.
   - Official optional skills can still be first-party and trusted; they just should not be seeded into every profile automatically.

3. Treat Apple skills as optional, not default-bundled
   - `apple-notes`, `apple-reminders`, `findmy`, and `imessage` are good candidates for opt-in official skills.
   - They could remain discoverable in the Skills Hub and/or an official optional-skills catalog.

4. Add an explicit allowlist for what gets seeded by `sync_skills()`
   - Instead of syncing every repo-bundled skill into every profile, seed only an approved core list.
   - Everything outside that allowlist is discoverable/installable but not auto-seeded.

5. Consider a migration path
   - Existing installs can keep already-synced skills unless the user removes them.
   - New installs and new profiles should get the slimmer default set.

This would preserve discoverability while making Hermes feel more modular and reducing the amount of bundled content that silently propagates across profiles.

### Alternatives Considered

Alternatives considered:

- Keep all skills bundled, but improve deletion/sync semantics.
[...]

PR #11894, opened 2026-04-18
feat(skills): add auto_sync_bundled config to disable bundled skill auto-copy

JorkeyLiu:
## What

Add a new config option: `skills.auto_sync_bundled` (default: `true`).

When set to `false`, Hermes will stop automatically copying new bundled skills into the user's `~/.hermes/skills/` directory during sync/update.

This solves the case where users intentionally keep a minimal local skills set and don't want newly added official skills to silently appear after upstream updates.

Existing bundled skills already tracked in the manifest still behave normally:
- unchanged skills stay unchanged
- updated bundled skills still update when the local copy is unmodified
- deleted skills remain respected

## Why

The existing manifest system already respects user deletions, but it cannot prevent newly introduced bundled skills from being auto-installed.

This config flag adds a clean opt-out for that behavior.

## Implementation

- Add `skills.auto_sync_bundled` to config defaults
- Gate the new bundled skill copy path in `tools/skills_sync.py`
- Preserve backward-compatible default behavior (`true`)
- Parse bool-like string values safely (`'false'`, `off`, `0`, etc.)
- Ensure `reset_bundled_skill(..., restore=True)` explicitly bypasses the suppression and still works even when auto-sync is disabled
- Surface `skipped_auto` in CLI update/sync output instead of incorrectly saying "Skills are up to date"

### Disable auto-sync for new bundled skills

```yaml
skills:
  auto_sync_bundled: false
```

Then run Hermes update / sync:
- newly introduced bundled skills should NOT be copied
- CLI should report skipped bundled skills

### Re-enable

Set it back to `true` (or remove the key):
- previously skipped bundled skills should be picked up on the next sync

### Explicit reset/restore still works

Even with `auto_sync_bundled: false`, `reset_bundled_skill(..., restore=True)` should still restore the selected bundled skill.

## Tests

Passed:
- `tests/tools/test_skills_sync.py`
- `tests/hermes_cli/test_config.py`
- `tests/tools/test_browser_camofox_state.py`

Key coverage added:
- auto-sync disabled skips new bundled skills
- existing tracked bundled skills still update normally
- quoted string `'false'` is parsed correctly
- explicit restore bypasses auto-sync suppression

## Platforms

- All (Python-only config + sync behavior)

Issue #65828, opened 2026-07-16
[Bug]: dashboard console skills commands act on the wrong profile — tools/skills_sync.py binds SKILLS_DIR at import and _profile_scope doesn't retarget it


Fe2-O3:
`hermes_cli/web_server.py:13566` `_profile_scope()` documents this exact bug class, and fixes it for two modules:

> 2. `tools.skills_tool` and `tools.skill_manager_tool` bind `SKILLS_DIR` at import time, so the override CANNOT reach them. Like `_call_cron_for_profile` does for cron's module globals, temporarily retarget both under a lock and restore them immediately after.

`tools/skills_sync.py:39-41` has the identical import-time binding and is **not** retargeted.

## The chain

1. `_execute_console_line()` (`web_server.py:15489`) wraps dashboard console dispatch in `with _profile_scope(profile):`, which calls `set_hermes_home_override(profile_dir)` and retargets only `skills_tool` + `skill_manager_tool`.
2. `("skills","reset")`, `("skills","diff")`, `("skills","list-modified")`, `("skills","opt-in")`, `("skills","opt-out")`, `("skills","repair-official")` are console-allowed (`console_engine.py:194-208`) and dispatch **in-process** — unlike hub `install`/`update`/`uninstall`, which correctly go through a fresh profile-scoped subprocess (`web_server.py:12501`).
3. `hermes_cli/skills_hub.py` lazily imports from `tools.skills_sync`: `reset_bundled_skill` (:1152), `list_user_modified_bundled_skills` (:1195), `diff_bundled_skill` (:1222), `sync_skills`/`set_bundled_skills_opt_out` (:1337), `restore_official_optional_skill` (:1365).
4. Those functions read the frozen module constants: `MANIFEST_FILE` (:104), `SKILLS_DIR` (:307, :340, :744), `HERMES_HOME` (:496, :1052).

So `skills_sync` binds to whichever profile first triggered a console skills command, and every later console skills op under a different profile resolves against that stale home.

## Why this one bites harder than its two siblings

`reset_bundled_skill()` **deletes**, and its safety boundary is the frozen constant. `skills_sync.py:744` does `skills_root = SKILLS_DIR.resolve()`, which backs the strict-child check whose own comment says it exists so a collapsed `dest` can't "wipe every installed skill". Under a mismatched profile that guard is computed against the wrong root.

## Precedent — this is the third module in the same lineage

[...]

PR #97308, opened 2026-08-28
feat(skills): opt-in bundled-skills whitelist (bundled_whitelist / bundled_enabled)

### LINKED ISSUE #35511: [Feature]: Bundled skills whitelist option
### Problem or Use Case

I run my agent with a local model and it's quite context constrained. Therefore i disabled almost all skills to save context. Currently skills are enabled/disabled via blacklist. 

On updates newly added bundled skills are automagically seeded into agents skills dir, and since they aren't blacklisted they are enabled by default. So my updates go: update -> start session -> notice random new skills -> close session -> list and disable new skills -> start session

Ultimately the desired change is not having skills just appear.

### Proposed Solution

bundled_whitelist config option (off by default, so existing users are unaffected) that, when set, uses skills.bundled_enabled: []  to filter bundled skills in addition to the skills.disabled pass. 

```
skills:
  bundled_whitelist: true
  bundled_enabled:
    - airtable
    - apple-notes
  disabled:
    - unwanted-user-skill
```
bundled_whitelist: false (default) — no filtering, existing behavior.
bundled_whitelist: true with an empty list — all bundled skills blocked.
bundled_whitelist: true with entries — filters to those

User/agent skills are unaffected by the whitelist and remain manageable via disabled as usual.

### Alternatives Considered

skills.new_bundled_default: disabled - config option to automatically add newly added bundled skills to the disabled list

The whitelist approach is better for users who only use a subset of bundled skills because a blacklist grows with every update while a whitelist stays stable.

### Feature Type

Configuration option

### Scope

None

### Contribution

### Debug Report (optional)

```shell

```

DavidMetcalfe:
## Summary

Add an opt-in runtime whitelist for bundled skills, implementing #35511. When `skills.bundled_whitelist: true`, bundled skills **not** listed in `skills.bundled_enabled` are treated as disabled — in addition to the existing `skills.disabled` pass. Default is off, so behavior for existing users is byte-identical.

```yaml
skills:
  bundled_whitelist: true
  bundled_enabled:
    - airtable
    - hermes-agent
  disabled:
    - unwanted-user-skill
```

- `bundled_whitelist: false` (default) — no filtering, existing behavior
- `true` + empty list — all bundled skills blocked
- `true` + entries — only those bundled skills active
- User/agent-authored skills are never affected by the whitelist
- Essential skills (`hermes-agent`) are never blocked, same as the disabled list

## Why

The issue author runs a context-constrained local model and disables almost all skills. On every update, newly-added bundled skills are auto-seeded into the skills dir and — not being blacklisted — appear enabled. The user's update loop today is: update → start session → notice random new skills → close session → disable them → start session. A whitelist stays stable as bundled skills grow; a blacklist grows with every update.

## Implementation

Bundled skills are copied into `~/.hermes/skills/` at sync time, so bundled origin at runtime is identified via the profile-scoped `.bundled_manifest` (`name:hash` lines). The whitelist is folded into the *effective disabled set* at all three existing disable-resolution layers, so list filtering, per-skill view gates, and cache invalidation all come for free:

- **`agent/skill_utils.py`** — new `_bundled_whitelist_blocked()` helper (manifest parsed via `get_hermes_home()` for call-time profile safety, no `tools.*` import per layering) unioned into `get_disabled_skill_names()`. This automatically feeds the `_find_all_skills` cache signature, so toggling the whitelist or editing `bundled_enabled` invalidates the skill-scan cache.
- **`tools/skills_tool.py`** — `_is_skill_disabled()` applies the same filter (per-skill view gates at plugin listing, plugin view, and skill_view).
- **`hermes_cli/skills_config.py`** — the `hermes skills` TUI shows whitelist-blocked skills as off, but the save path subtracts them before persisting `skills.disabled`, so toggling the whitelist off later never leaves stale disabled entries. Toggling only whitelist-blocked skills prints a pointer to `skills.bundled_enabled` instead of silently saving nothing.
- **`hermes_cli/config_defaults.py`** — `bundled_whitelist: false`, `bundled_enabled: []` registered with full semantics docs.
- Slash-command listings (`hermes_cli/commands.py`) resolve through `get_disabled_skill_names()` and inherit the filter automatically.

**Edge semantics:** if the whitelist is on but the manifest is missing (bundled skills never synced), nothing is filtered — fail-open, documented in `config_defaults.py` — so enabling it on a fresh install cannot hide every skill.

## Testing

- 42 new tests in `tests/tools/test_skills_bundled_whitelist.py` + `tests/hermes_cli/test_skills_config.py`: off-by-default, empty-list-blocks-all, entries-filter, user-skill immunity, essential exemption, `_is_skill_disabled` parity, TUI save-path non-pollution, cache-signature invalidation, plus regression tests for review findings (`skills: null` guard, invalid-UTF-8 manifest, masked-count math).
- Full run: `162 passed` across the whitelist, skills-config, skills-sync, skills-tool, discovery-cache, profile-scope, and skill_utils/bundles suites.

## Review

Two rounds of cross-vendor review (Gemini 3.7 Flash High + GPT-OSS 120B Medium) on the full diff, then the fix delta. Round-1 fixes: `skills: null` AttributeError guard, `UnicodeDecodeError` tolerance on manifest read, TUI masked-count save path, dead helper removed. Round 2 confirmed all fixes and surfaced one UX regression (restored the zero-toggle early-return). Round-1 false positives rejected after source verification: manifest-name case asymmetry (the normalization helper does not lowercase) and raw-config cache staleness (mtime-keyed cache, config passed as parameter).

Closes #35511

See also: #19986 (broader optional-bundled-skills discussion), #11894 / #29013 (sync-time auto-copy suppression — a complementary layer that would additionally prevent new bundled skills from being *copied* at update time; this PR handles the runtime filter independently and works regardless of sync behavior).

### DIFF
--- a/agent/skill_utils.py
+++ b/agent/skill_utils.py
@@ -469,11 +469,11 @@ def _read_bundled_manifest_names() -> Set[str]:
                 # v1 format: plain name
                 names.add(line)
         return names
-    except (OSError, IOError):
+    except (OSError, IOError, UnicodeDecodeError):
         return set()
 
 
-def _bundled_whitelist_blocked(skills_cfg: dict) -> Set[str]:
+def _bundled_whitelist_blocked(skills_cfg: dict | None) -> Set[str]:
     """Return skill names that the bundled-skills whitelist blocks.
 
     When ``bundled_whitelist`` is truthy in *skills_cfg*, any bundled-provenance
@@ -484,7 +484,7 @@ def _bundled_whitelist_blocked(skills_cfg: dict) -> Set[str]:
     Essential skills (``ESSENTIAL_SKILLS``) are never blocked even when they
     are bundled and not in the enabled list.
     """
-    if not skills_cfg.get("bundled_whitelist"):
+    if not isinstance(skills_cfg, dict) or not skills_cfg.get("bundled_whitelist"):
         return set()
 
     bundled_names = _read_bundled_manifest_names()
--- a/agent/skill_utils.py
+++ b/agent/skill_utils.py
@@ -443,6 +443,59 @@ def _load_raw_config() -> Dict[str, Any]:
 ESSENTIAL_SKILLS: frozenset = frozenset({"hermes-agent"})
 
 
+def _read_bundled_manifest_names() -> Set[str]:
+    """Read the set of skill names from the active profile's bundled-skills manifest.
+
+    Parses the same 'name:hash' line format as ``tools.skills_sync._read_manifest``.
+    Returns an empty set when the manifest is missing, unreadable, or the profile
+    has no bundled skills. Does NOT import tools.skills_sync to avoid layering
+    violations (agent/ is lower than tools/).
+    """
+    from hermes_constants import get_hermes_home
+
+    manifest_path = get_hermes_home() / "skills" / ".bundled_manifest"
+    if not manifest_path.exists():
+        return set()
+    try:
+        names: Set[str] = set()
+        for line in manifest_path.read_text(encoding="utf-8").splitlines():
+            line = line.strip()
+            if not line:
+                continue
+            if ":" in line:
+                name, _, _ = line.partition(":")
+                names.add(name.strip())
+            else:
+                # v1 format: plain name
+                names.add(line)
+        return names
+    except (OSError, IOError):
+        return set()
+
+
+def _bundled_whitelist_blocked(skills_cfg: dict) -> Set[str]:
+    """Return skill names that the bundled-skills whitelist blocks.
+
+    When ``bundled_whitelist`` is truthy in *skills_cfg*, any bundled-provenance
+    skill whose name is NOT in ``bundled_enabled`` is treated as effectively
+    disabled (blocked). When the whitelist is off (default) or the whitelist
+    config is absent, returns an empty set — zero behavior change.
+
+    Essential skills (``ESSENTIAL_SKILLS``) are never blocked even when they
+    are bundled and not in the enabled list.
+    """
+    if not skills_cfg.get("bundled_whitelist"):
+        return set()
+
+    bundled_names = _read_bundled_manifest_names()
+    if not bundled_names:
+        return set()
+
+    enabled = _normalize_string_set(skills_cfg.get("bundled_enabled"))
+    blocked = bundled_names - enabled
+    return blocked - ESSENTIAL_SKILLS
+
+
 def get_disabled_skill_names(platform: str | None = None) -> Set[str]:
     """Read disabled skill names from config.yaml.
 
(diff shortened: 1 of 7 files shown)


DavidMetcalfe:
CI is green on the latest commit — all required checks pass, including the Python test suite, lints, OS-specific tests (macOS/Windows), and the supply-chain/OSV scans. The remaining `skipping` entries are path-filtered jobs (Rust, JS/TS, Docker lint, installer tests, Docs Site) that don't apply to this diff.

Ready for review.

AureliusClaw:
Reviewed this independently (code + tests) — genuinely well done. This directly fixes the failure mode in #35511 and does it the right way.

What stands out:

- **Composition, not replacement.** The whitelist is an opt-in *additional* filter unioned with the existing `skills.disabled` denylist rather than special-casing or replacing it. Defaults are off (`bundled_whitelist: false` / `bundled_enabled: []`), so nobody who doesn't opt in sees any behavior change. That's the correct contract for a capability-hiding feature.
- **Fail-open is the right default.** Missing/unreadable `.bundled_manifest` filters nothing, so enabling the whitelist on an install that never synced bundled skills can't accidentally hide everything. Documented in `config_defaults.py` in exactly the words the code implements.
- **Essential skills are protected in the right place.** Subtracting `ESSENTIAL_SKILLS` inside `_bundled_whitelist_blocked()` means all three consumers inherit it — including the `_is_skill_disabled()` early-return path in `skills_tool.py` that has no trailing subtraction of its own. Easy to get wrong; this is deliberate and correct.
- **The save-time subtraction in `skills_command()` is the subtle bit most implementations miss.** Not persisting whitelist-blocked names into `skills.disabled` means toggling the whitelist off later won't leave a trail of stale denials, and surfacing "+N whitelist-blocked" with a pointer to `bundled_enabled` makes the interaction legible instead of confusing.
- **Provenance from the canonical manifest at call time.** `get_hermes_home()` re-resolves per call, inheriting the multi-profile fix from #65828 — so under a non-default active profile (skills root at `~/.hermes/profiles/<name>/skills/`) it reads that profile's manifest, not the launch profile's. I verified this holds, and it's the scenario #35511 actually needs.
- **Reuse across all three enforcement sites** (`get_disabled_skill_names`, `_is_skill_disabled`, CLI `get_disabled_skills`) with an explicit parity test between the first two — exactly the drift risk given three near-identical composition sites.

I checked out the branch and ran the suite: all 42 new tests pass, plus the surrounding skill/sync tests (~180), lint clean. The test file is proportionate for a feature that can hide capabilities and covers the cases that matter (off/absent/empty/partial, essential-skill and user-skill negatives, both manifest formats, missing/unreadable manifest, and the parity + cache-invalidation paths).

I'm a maintainer of a Hermes profile that runs a context-constrained model and disables most bundled skills — this PR removes the one thing that makes that setup brittle, so I have a concrete second use case riding on it. It's mergeable with CI green and would be great to see land. Thanks for taking this on.

PR #96038, opened 2026-08-27
fix(state): fail closed on unscoped corruption

diatche:
## Summary

- require FTS-specific evidence before entering runtime FTS repair/fail-open
- accept `SQLITE_CORRUPT_VTAB` for older SQLite builds whose text is only `database disk image is malformed`
- propagate bare `SQLITE_CORRUPT` and contradictory known result codes without rebuilding FTS, dropping FTS triggers, or marking the index stale

## Why

`_is_fts_write_corruption_error()` currently treats every malformed-image error as FTS corruption. That assumption is unsafe: the same text also reports structural damage to canonical tables or the database header. Routing an unscoped structural failure into live FTS mutation can obscure the real failure and compound recovery.

This preserves the compatibility reason behind 987064caa4 (older SQLite may use generic text for corrupt FTS) by using SQLite's extended result code rather than treating the message alone as proof. It also restores the FTS-class gate expected by #89073 without duplicating that PR's broader offline-repair design.

Known SQLite error codes take precedence over text: `SQLITE_CORRUPT_VTAB` is accepted, while a contradictory code such as `SQLITE_CONSTRAINT_TRIGGER` fails closed even if its message contains FTS-looking corruption words. The narrow text fallback is used only when no error code is available.

## Tests

- generic malformed-image error without an extended FTS code is rejected
- base `SQLITE_CORRUPT` is rejected
- `SQLITE_CORRUPT_VTAB` is accepted even with generic text
- explicit `fts5: corrupt structure ...` remains accepted when no code is available
- FTS-looking text paired with `SQLITE_CONSTRAINT_TRIGGER` is rejected
- structural and contradictory-code failures propagate through `_execute_write()` without FTS rebuild, stale marker, or trigger removal

Local validation:

```text
tests/state + tests/hermes_state: 271 passed
ruff check: passed
ruff format (changed ranges): passed
py_compile: passed
git diff --check: passed
```

PR #69042, opened 2026-07-22
fix(gateway): per-profile session isolation for shared-WhatsApp multiplex

lsshawn:
## Problem

Running `gateway.multiplex_profiles` with a **single shared WhatsApp (Baileys) account** routed to multiple profiles via `gateway.profile_routes`, inbound messages were served by the **default** profile's config/skills/memory/session-DB even when a `profile_route` matched — leaking default-profile data (memory, `session_search` history, filesystem) into routed replies. Three related gaps:

### 1. Secondary profiles tried to start their own WhatsApp adapter
`_start_one_profile_adapters` attempted a WhatsApp connection per profile. With one shared account, secondaries have no creds → retry-loop that **stalls gateway startup**. WhatsApp is a single-shared-connection platform like `RELAY`, so we now skip it for secondaries (mirrors the existing `RELAY` skip). The default profile owns the one connection; `profile_routes` routes groups.

### 2. Session-key namespace ignored `profile_routes`
`SessionStore._resolve_profile_for_key` fell back to the active profile (default) when `source.profile` was unset — even when a `profile_route` matched — so routed sessions landed in `agent:main` + the default `state.db`. Now consults `gateway.profile_routes` and stamps `source.profile` so the key namespace agrees with routing.

### 3. `SessionStore._db` + inbound session creation were default-bound
`SessionStore` held a single default-home `SessionDB`; every session's history/search/persistence resolved to default's `state.db` regardless of profile. Made `_db` a per-home-cached property (resolves `get_hermes_home()/state.db` under the active ContextVar scope), and wrapped inbound `get_or_create_session` in `_profile_runtime_scope` so the session is physically created in the routed profile's `state.db`.

## Result
A WhatsApp group routed to profile X now runs fully as X — its own skills, memory, and `state.db` — and `session_search` no longer surfaces another profile's history.

## Safety
All new behavior is gated on `multiplex_profiles`. Multiplex-**off** behavior is byte-identical.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #66887, opened 2026-07-18
Multiplexed gateway: secondary-profile Telegram sessions persist in the default profile's state.db / sessions.json (creation AND recovery paths) — config/SOUL/memory scope correctly, storage does not


marknisk:
*(Apologies for the brief placeholder content on first submission — tooling slip on our side, fixed within minutes.)*

Deployment: v2026.7.7.2 (0.18.2), Docker, s6-supervised gateway, `multiplex_profiles: true`, one secondary profile (`home-ops`) owning a Telegram DM adapter. Container env carries a global `HERMES_HOME=/opt/data` (the default scope); the profile lives at `/opt/data/profiles/home-ops`.

Observed: every gateway turn for the home-ops-owned Telegram adapter runs with the CORRECT profile scope for config, model/provider, SOUL, and the Honcho memory peer (the 0.18.x scope-aware layers work) — but session persistence and logging land in the DEFAULT store:

- session rows + messages are written to `/opt/data/state.db`; the profile's `state.db` has no trace of them,
- the session registry entry (key `agent:home-ops:telegram:dm:<chat>`) lives in `/opt/data/sessions/sessions.json`, which re-seeds `gateway_routing` at boot,
- turn logs go to the default profile's `agent.log`.

This is not stale state. We removed the registry entry, the `gateway_routing` row, and archived the old session, then restarted: a FRESH session created by the next inbound Telegram message was again persisted in `/opt/data/state.db` (correct SOUL/model/peer, wrong store). The recovery path has the same bias — after a restart marked the original (correctly profile-scoped, created 0.18.2 day one) session ended, the "ended in state.db but still live in sessions.json; recovering" path recreated it in the default store.

The CLI path is fine: `hermes -p home-ops chat` persists to the profile store (confirmed while corroborating the closure of #63978). The seam is gateway-only: session-store resolution appears to use the process-global `HERMES_HOME` rather than the adapter's owning profile, unlike the config path (`_gateway_config_home()`).

Impact: the profile's session history/compaction/billing/FTS live in the wrong database, and a deployment using profiles for privilege separation gets secondary-profile conversation content mingled into the default store. Suggestion: resolve the session store (and log sink) from the adapter's owning profile, the same way 0.18.x already resolves config.

alt-glitch:
> *This was generated by AI during triage.*

Related to the multiplex-profile cluster, including #57492, but this report targets profile-scoped session-store/routing persistence rather than adapter selection or reply delivery.

[...]

Issue #54878, opened 2026-06-29
[Bug]: idle agent cache eviction ends session in DB but leaves sessions.json stale → messages silently dropped until restart


agt-user:
## Bug Description

When an AIAgent is evicted from the in-memory agent cache due to idle TTL expiry (`_AGENT_CACHE_IDLE_TTL_SECS = 3600s`), `end_session(session_id, "agent_close")` is written to `state.db` — but `sessions.json` is **not updated**. The stale `session_key → session_id` entry remains, pointing to a now-ended session.

All subsequent messages on that session_key are **silently dropped** — no log output, no error, no visible failure — until the gateway is restarted (which triggers `_prune_stale_sessions_locked()` on startup).

This is distinct from FM9 (#52804, fixed locally by #52808 via startup pruning) which requires a **gateway crash**. In this variant, the gateway is fully alive and healthy (ESTAB sockets, normal CPU/memory, `active (running)` in systemd), and the session ended gracefully via idle eviction — not a crash.

## Steps to Reproduce

1. Start gateway with default config (`_AGENT_CACHE_IDLE_TTL_SECS = 3600`)
2. Have a conversation (any length)
3. Wait 60+ minutes without sending any messages to that session
4. `_sweep_idle_cached_agents` fires → `_evict_cached_agent(session_key)` → `_release_evicted_agent_soft(agent)` → `agent.release_clients()` → `end_session(session_id, "agent_close")` written to DB
5. `sessions.json` still maps the session_key to the now-ended session_id
6. Send any message
7. **Message silently dropped** — no response, no journalctl output, no error

## Field Incident (2026-06-29)

```
# Timeline
10:49:07  session 20260629_104844_fe3a0e87 started (Luna DM)
10:49:52  Last assistant message sent

# ~50 min later — idle eviction fires (AIAgent evicted from cache)
# end_session(..., "agent_close") written to state.db
# sessions.json NOT cleared

11:39:22  User sends message to Luna
# Telegram delivers to Luna's long-poll (pending_update_count=0 confirms delivery)
# sessions.json lookup → 20260629_104844_fe3a0e87 → end_reason='agent_close'
# Message silently dropped — no log, no journal, no response

11:48:xx  User notices Luna is unresponsive
11:54:14  Manual SIGUSR1 restart
11:54:24  Startup pruning catches it:
          WARNING gateway.session: pruning stale sessions.json entry
[...]

PR #97309, opened 2026-08-28
fix(gateway): resolve session storage from the key's profile, not ambient scope

caya8205-2:
## What does this PR do?

Resolves a `SessionStore` operation's `state.db` from the profile encoded in the session key instead of from the ambient `HERMES_HOME`, so work that runs outside `_profile_runtime_scope` stops reading and writing the wrong profile's store.

#88734 made `SessionStore._db` follow the ambient scope. That is correct for the inbound message path — that path installs the scope. Nothing else does. `_session_expiry_watcher` (`gateway/run.py`) walks the single process-wide `_entries` dict, which holds **every** profile's keys, and finalizes expired sessions with no scope installed, so `_db` resolves the **root** store for rows that live under `profiles/<name>/state.db`.

The scoped inbound path and the unscoped background path then maintain two copies of the same logical session, and their `end_reason` values drift apart independently. Once they disagree, the `#54878` stale-routing guard reads one copy while the routing index points at the other, and a live conversation is dropped and recreated — silently, because that branch only sets `was_auto_reset` when a reset policy also fired.

Field evidence from a live multiplex install (v0.20.4, two named profiles on one Telegram bot), read-only `sqlite3` across the stores:

| session id | root `state.db` | `profiles/erp-u2-c1/state.db` |
|---|---|---|
| `20260814_234313_6017bb94` | `end_reason=None` (alive) | `agent_close` (ended) |
| `20260822_225807_11876f35` | `session_switch` (ended) | `end_reason=None` (alive) |

The disagreement runs in **both directions**, which rules out a single mis-scoped writer: these are two independently-scoped write paths racing on one session. Both ids are from the cascading self-heal I reported in #66887.

## Related Issue

Part of #66887. This PR covers the per-session half — which store a row is read from and written to. The routing-index half (`_routing_scope()` and the `sessions.json` mirror are still pinned to one frozen `sessions_dir` while the handle moves) is a follow-up; I did not want to mix the two.

## Type of Change

Bug fix — data loss (dropped conversation history).

## Changes Made

`gateway/session.py`

- `_profile_home_for_key()` — the owning profile's `HERMES_HOME`, memoized per profile name. Returns `None` when multiplexing is off, when the key carries the legacy `agent:main` namespace, or when the profile has no live directory.
- `_db_for_key()` / `_db_for_session_id()` — the store that owns a routing key, or a session id (recovered from the in-memory index; deliberately lock-free, since several callers already hold `_lock`, and a miss just falls back to the previous ambient behavior).
- `_open_session_db_for_active_scope(db_path=None)` — accepts an explicit path. The default branch still calls `SessionDB()` exactly as before.
- 40 `self._db` uses across 14 methods now resolve through those helpers: `_get_or_create_session_impl`, `reset_session`, `switch_session`, `rewind_session`, `_recover_session_from_db`, `_find_gateway_session_row`, `_record_gateway_session_peer`, `_is_session_ended_in_db`, `set_expiry_finalized`, `load_transcript`, `rewrite_transcript`, `append_to_transcript`, `has_platform_message_id`, `_compression_tip_for_session_id`.
- Genuinely store-wide callers (`has_any_sessions`, `_rebuild_fts_once`) deliberately keep ambient resolution.

`tests/gateway/test_multiplex_session_db_profile_scope.py` — five tests plus a module-docstring paragraph.

**No signature changed and no existing test was modified.**

Two properties worth calling out, since they were the objections raised on #69042:

- **Single-profile installs are untouched.** A key without a named profile resolves exactly where it did before these helpers existed. `test_default_namespace_keeps_ambient_resolution` pins that.
- **The live-DB guard still covers the new path.** The explicit-path branch goes through `SessionDB.__init__` → `_ensure_test_isolation`, and `_is_production_state_db` already understands the `profiles/<name>/state.db` layout.

## How to Test

```
pytest tests/gateway/test_multiplex_session_db_profile_scope.py -q
```

Against this branch: **16 passed** (11 pre-existing + 5 new).

Against `main`, the new tests behave as follows — the distinction matters, so I'll be precise rather than claim all five reproduce:

| test | on `main` |
|---|---|
| `test_scoped_inbound_turn_lands_in_profile_store` | **passes** — deliberate control; the scoped path was already fixed by #88734 |
| `test_unscoped_background_finalize_reaches_the_key_owner_store` | **fails on wrong data** — `expiry_finalized` is still `0` on the profile's row |
| `test_unscoped_staleness_check_reads_the_key_owner_store` | **fails on wrong data** — a live session is reported as ended |
| `test_default_namespace_keeps_ambient_resolution` | fails with `AttributeError` (guardrail; references the new helpers) |
| `test_pinned_handle_still_wins_over_key_resolution` | fails with `AttributeError` (guardrail; references the new helpers) |

The two middle rows are the actual reproduction: they fail with incorrect values, not merely a missing attribute. The control passing on `main` is what isolates the variable — the only difference between it and the row below is whether a scope is installed.

Wider run, 6 storage/multiplex suites — `test_multiplex_session_db_profile_scope`, `test_session_store_runtime_stale_guard`, `test_session_store_stale_prune`, `test_multiplex_phase0`, `test_session_store_lock_io`, `test_session_store_expiry_finalized`: **56 passed**.

`ruff check` clean. `scripts/check-windows-footguns.py` clean.

Tested on: Debian 13 container (Python 3.11, `uv sync --locked --extra all --extra dev`), Docker on Windows 10 Pro 19045.

### Note on a full `tests/gateway/` run

A single-process `pytest tests/gateway/` gives **9 failed, 6528 passed, 45 skipped, 2 xfailed**. I checked each failure against pristine `main` rather than assuming:

- `test_session_hygiene`, `test_session_store_prune::test_session_store_default_db_uses_runtime_hermes_home`, `test_turn_lease` — **already fail on pristine `main`** in this environment, unrelated to this change.
- The other six (`test_discord_send` ×3, `test_send_multiple_images`, `test_teams_dotenv_isolation`, `test_telegram_polling_health_confirmation`) **pass in isolation on both pristine `main` and this branch** — cross-file pollution from running one process, which is precisely what `scripts/run_tests.sh` per-file isolation exists to avoid. None of them touch session storage.

---

## Update after review

Head `373bb3e227b3478d4150738827ad16a8b73e6d9a`, recomposed on current `main`. Three review rounds folded in:

1. The profile-home memo cached the *miss*, pinning a profile provisioned after startup to the ambient store for the process lifetime. Hits only are memoized now.
2. `_profile_home_for_key()` collapsed three states into one `None`. Ownership is tri-state via `_named_profile_for_key()`: no named owner → ambient; named owner + home → that profile; named owner, unresolvable → `None` and a warning, never root.
3. The compression continuation was written to the child before its reroute and `_entries` update published, so the child id resolved to the ambient store. A private `_session_owner_hints` map proves ownership before the mutation; publication ordering is unchanged.

Test count is now 8 in `tests/gateway/test_multiplex_session_db_profile_scope.py`, and 126 pass across the storage/multiplex/compression suites.

### Known interaction: `SessionDB._execute_write` recovery classification

Flagged in review and recorded here rather than left implicit. #97940 reports that a bare `SQLITE_CORRUPT` from a canonical B-tree is classified as FTS-only, after which `_try_runtime_fts_rebuild()` / `_enter_fts_fail_open()` may write `state_meta`, drop FTS triggers, change the handle's mode and retry the mutation.

This PR does not introduce, modify or newly reach that machinery: unscoped background writes already ran through `_execute_write` before this change — against the **root** store, because ambient resolution sent them there. What does change is the spread: writes that previously concentrated on one database now land across each owning profile's database.

#96038 carries the fail-closed classification. This PR is deliberately not gated on it, so that a data-loss repair is not held behind a separate unmerged change, but the interaction is real and the cross-profile corruption regression belongs with that fix.


Enough1122:

**Overall:** Critical correctness fix for multiplex gateways that resolves session storage from the profile encoded in the session key. Stops unscoped background work (notably `_session_expiry_watcher` walking every profile's keys) from writing expiry/muting state to the root store and drifting replicas apart, which triggered silent self-heal recreation of live conversations (#66887).

**What it does**
- `gateway/session.py:322` adds `_profile_home_cache: Dict[str, Optional[Path]]` memo, refactors `_open_session_db_for_active_scope(db_path=None)` to accept explicit `db_path`, introduces `_profile_home_for_key(session_key)` (multiplex-gated, `profile_exists`/`get_profile_dir` with `cache`, `None` → ambient for `default`/single-profile legacy `agent:main`), `_db_for_key(session_key)` (pinned check → home → `_open_session_db_for_active_scope(home/state.db)` → fallback to ambient on error) and `_db_for_session_id(session_id)` (recover key via `_entries` linear scan, lock-free, fallback to ambient).
- Replaces direct `_db` reads/writes in expiry finalization and `_is_session_ended_in_db` paths to use `_db_for_key`/`_db_for_session_id`, ensuring background finalize reads/writes the owning profile's `state.db`.
- Tests `tests/gateway/test_multiplex_session_db_profile_scope.py:179` cover scoped inbound still profile-scoped, unscoped background finalize reaches key-owner store, staleness check reads key-owner store (alive vs ended), and legacy default namespace keeps ambient resolution; plus multiplex profile topic and contributors note.

**Non-blocking notes**
- `_db_for_session_id` linear scan over `list(self._entries.values())` — called on staleness check; acceptable as entries count modest, but note O(n) per check under large multiplex — cache could help if scaled.
- Memo `_profile_home_cache` never invalidated on profile deletion — profile existence checked once; if profile removed mid-process, stale `None` cached vs live dir; rare but consider eviction on `profile_exists` false.
- Keep routing-index half of #66887 (`_routing_scope`/`sessions.json` mirror) as separate follow-up — well scoped as noted.

caya8205-2:
Thanks — both notes are fair. The second one is a real bug, so I've pushed a fix (`80157e7`).

**Memo not invalidated.** It's sharper than deletion, actually: the *miss* was cached, so a profile that appears **after** the gateway starts stays pinned to the ambient store for the life of the process — precisely the failure this helper exists to prevent. That isn't hypothetical: an enrollment bridge can provision `profiles/<name>/` at runtime, so a key is legitimately seen before its directory exists. `_profile_home_for_key` now memoizes only hits; a miss costs one `profile_exists()` stat and recurs only for profiles that genuinely don't exist, so the hot path for real profiles is still a dict hit. `test_profile_home_is_not_memoized_before_the_profile_exists` covers it — it fails without the change.

**O(n) scan in `_db_for_session_id`.** Leaving this one as-is, deliberately. The scan is bounded by routing-key count and runs once per call rather than per row, so it's tens of microseconds in front of a SQLite write that is orders of magnitude slower. And given the bug above came from a cache I added, I'd rather not add a second one keyed on `session_id`: the mapping looks immutable, but the legacy pre-workspace Slack fallback in `_query_recoverable_session` does rewrite a peer row onto the scoped key, so it would need a real invalidation story to buy a saving that doesn't show up in a profile. Happy to revisit if anyone has a deployment where it measures.

**Routing-index half** — agreed, keeping `_routing_scope()` and the `sessions.json` mirror as a separate follow-up once this one settles.

Suites after the fix: `test_multiplex_session_db_profile_scope`, `test_session_store_runtime_stale_guard`, `test_session_store_stale_prune`, `test_multiplex_phase0`, `test_session_store_lock_io`, `test_session_store_expiry_finalized` — 57 passed. `ruff` and `check-windows-footguns` clean.

andrexibiza (commented):
Reviewed exact head `80157e795ffc2f39dcc1a19eea7f561f01b2e29e` against the live multiplex-storage graph, the #66887 incident thread, merged #88734, the narrower historical #69042 objections, adjacent degraded-store work in #88804, and current main.

The core direction is right: moving session-row ownership from ambient `HERMES_HOME` to the profile already encoded in `agent:<profile>:…` closes the background-worker half that #88734 could not close. In particular, carrying `_db_for_key()` through expiry finalization, staleness checks, recovery, transcript mutation, reset/switch, and peer repair is the right shape. The field evidence showing opposite `end_reason` polarity in root vs profile stores is especially useful because it proves two independent writers rather than a single stale copy. The follow-up commit also correctly stopped memoizing a profile-home miss.

I found one P1 boundary that still recreates the exact wrong-store class during runtime enrollment:

### P1 — a named profile that is not live *yet* still falls through to the ambient/root DB

`_profile_home_for_key()` currently returns the same `None` for three semantically different states: multiplexing/default namespace, a named profile whose directory does not exist yet, and a profile-resolution exception. `_db_for_key()` then interprets every `None` as `return self._db`.

That matters because the new regression itself establishes that the enrollment bridge can legitimately expose a key before `profiles/<name>/` exists. On that first lookup, a key such as `agent:fitness:telegram:dm:1` therefore still resolves to the ambient root `state.db`; after enrollment creates the directory, the next lookup flips to `profiles/fitness/state.db`. That is two physical stores for one qualified session identity — precisely the split this PR is meant to eliminate. The same fail-open happens on a transient `profile_exists()` / `get_profile_dir()` error because the exception is also collapsed to `None`.

Please make ownership tri-state instead of using `None` for both “ambient is authoritative” and “named owner unavailable”:

- non-multiplex / legacy-default key → ambient DB, preserving current single-profile behavior;
- named profile + resolved home → that profile DB;
- named profile + missing/unresolvable home → no DB / defer or refuse the state mutation, never silently fall back to root.

The regression should exercise the real effect boundary, not just cache behavior: resolve a named key before its profile directory exists, attempt a representative write/read through `_db_for_key()`, and prove root `state.db` remains untouched; then create the profile and prove the same key lands only in the profile DB. Also cover the resolver-exception branch, because it currently has the same ambient fallback.

### Interlocks / landing edge

This is a continuation of #88734 rather than a duplicate: #88734 established per-active-scope handles and preserved the #88632 / `jackulau` lineage; this PR repairs the unscoped/background side by deriving ownership from the key. #69042 is historical architectural evidence for the still-unfinished routing-index half, not an alternate implementation. #88804 is complementary degraded-DB continuity work and should be reconciled at `load_transcript()` rather than treated as superseded.

The #66887 routing-index half remains real: `_entries` / `_loaded`, `_routing_scope()`, `gateway_routing`, and the `sessions.json` mirror are still one process-wide/frozen storage surface. I agree with keeping that out of this diff, but it must remain explicitly open and merge-ordered after this change; this PR should not be used as closure evidence for #66887 while that half can still disagree with the per-session owner.

Current topology also needs a fresh landing object. The reviewed head is **2 ahead / 133 behind** live `main@e387cbc0aa0fc89560bc14438762c7722663db05`. I checked the main-side history since the merge base and there are no newer commits touching `gateway/session.py`, so this is not presently a semantic collision claim — but exact-object receipts are non-hereditary. Exact-head CI `33253693491`, Docker `33253693108`, and Nix `33253693016` are all `action_required`; CI has zero jobs. The reported focused/local 57-pass result is useful development evidence, not the upstream landing receipt.

Once the named-owner-unavailable case fails closed, recompose on current main and rerun the focused multiplex/storage suites plus exact-head CI/Docker/Nix. The key-derived ownership work itself is a strong repair of a nasty live data-loss seam; the remaining bug is at exactly the last boundary where ambient state can still sneak back in. 🚀

caya8205-2:
Fixed and recomposed on current `main`. New exact head: `dd96d2d8ed14efe5f997acae4030928eece3f63b`.

**P1 — named owner that isn't live yet.** You're right, and it's the sharper version of the miss-caching bug from the previous round: I had stopped *memoizing* the miss but left the *behavior* on a miss unchanged. `_profile_home_for_key()` was collapsing three different states into one `None`, and `_db_for_key()` read every one of them as "ambient is authoritative".

Ownership is now tri-state, split out into `_named_profile_for_key()`:

- no named owner (multiplexing off, or the legacy `agent:main` namespace) → ambient DB, single-profile behavior untouched;
- named owner + resolved home → that profile's DB;
- named owner, unresolvable → `None` plus a `logger.warning`, never root.

Callers already treat a missing DB as "skip the mutation", which is the defer-don't-misroute behavior you asked for. The one path reached with an id the entry-point guard did not check is `_append_transcript_message` via the compression-child id, so it now raises explicitly and lets the existing retry queue hold the row, instead of depending on an `AttributeError` to arrive at the same place.

**Tests at the effect boundary, not the cache.** `test_named_owner_without_a_home_never_falls_back_to_root` resolves a named key before `profiles/latecomer/` exists, asserts root `state.db` is untouched, then creates the directory and asserts a write through the same key lands only in the profile store. `test_profile_resolution_failure_fails_closed` covers the exception branch by making `profile_exists` raise. Both fail against the previous two-state behavior by returning a live `SessionDB` where `None` is required — a behavioral failure, not a missing attribute.

**Landing object.** Rebased onto current `main`; no commits since the previous merge base touch `gateway/session.py`, so this was not a semantic collision. Focused suites after the rebase: 59 passed across `test_multiplex_session_db_profile_scope`, `test_session_store_runtime_stale_guard`, `test_session_store_stale_prune`, `test_multiplex_phase0`, `test_session_store_lock_io` and `test_session_store_expiry_finalized`; `ruff` and `check-windows-footguns` clean. Agreed that is development evidence and not the landing receipt — the exact-head CI/Docker/Nix runs sit at `action_required`, which I can't clear from here, so that part needs a maintainer.

One environment note for whoever reads the receipts: `tests/gateway/test_session_store_prune.py::test_session_store_default_db_uses_runtime_hermes_home` also fails on pristine `main` in my container (the conftest re-points `DEFAULT_DB_PATH`, which wins over the runtime `HERMES_HOME` the test sets). Unrelated to this change — I verified by reverting `gateway/session.py` and re-running.

**#66887 scope.** Agreed, and to be explicit: this PR should not be read as closing #66887. `_entries` / `_loaded`, `_routing_scope()`, `gateway_routing` and the `sessions.json` mirror are still one process-wide, frozen storage surface that can disagree with the per-session owner. I'll open that as a follow-up ordered after this one rather than folding it in here.

Thanks for the #88804 pointer — I'll reconcile at `load_transcript()` there rather than treating it as superseded.

andrexibiza (commented):
Re-reviewed exact head `dd96d2d8ed14efe5f997acae4030928eece3f63b` against the author's follow-up and live `main`.

The original named-owner-unavailable P1 is fixed. `_named_profile_for_key()` restores the required tri-state; missing and resolver-error named owners no longer collapse into ambient root, and the added tests exercise the storage effect boundary rather than only helper/cache state.

One P1 remains.

### P1 — the compression continuation loses its profile owner before the child write

`_append_to_transcript_serialized()` resolves the live `child_id` through the parent’s correctly selected profile DB, then calls `_append_transcript_message(child_id, msg)` **before** publishing either `_transcript_reroutes[parent] = child` or the `_entries` session-id update. At that moment no routing entry points at `child_id`, so `_db_for_session_id(child_id)` misses its `_entries` scan and executes its documented fallback, `_db_for_key(None)`: the ambient DB. The new `if _db is None` guard therefore does not fire because root is still a live `SessionDB` handle.

That makes the mutation target root rather than the already-proven owner. With no child row in root, the append is rejected and the pending queue/reroute cannot advance because publication occurs only after the child write succeeds. If root contains a split-brain copy—the exact incident class this PR is repairing—the message can instead be written cross-profile before routing advances.

Carry the already-resolved parent owner DB/key through the continuation append, or establish a private fail-closed `session_id -> owner` proof before mutation. Do not publish the public reroute early; the current queue-migration-before-publication ordering is load-bearing for backlog order.

Add a real multiplex regression with an ended parent and live compression child existing only in `profiles/fitness/state.db`, clear the active profile/home scope, append to the parent, and prove all four effects: the message lands on the child in the profile DB; the pending queue drains; the reroute/current entry advances; root `state.db` remains untouched.

Topology/landing gate: this head is now 36 commits behind `main@3f36c87e1ebdfbf7d14a88229dc9be222c12ea89`. The intervening changes do not touch `gateway/session.py`, but `gateway/run.py` moved in the adjacent execution surface, so recompose on current main after the fix. Exact-head CI `33256914870`, Docker `33256914235`, and Nix `33256914233` are all `action_required`; there is still no green exact-head landing receipt.

andrexibiza (commented):
**P1 — the proof covers profile selection, but the new write surface can execute broader, unsafe recovery mutations on that selected profile.**

The owner-resolution direction is correct, but the current base still has #97940: a bare `SQLITE_CORRUPT` from a canonical B-tree is classified as FTS-only. Before the requested operation returns, `SessionDB._execute_write()` may run `_try_runtime_fts_rebuild()` and then `_enter_fts_fail_open()`, which can write `state_meta`, drop the FTS trigger set, change the handle’s in-memory mode, and retry the canonical mutation under the false premise that canonical rows are preserved.

This PR newly sends unscoped background mutations—expiry finalization, reset/end/create paths, transcript writes, and related work—into each named profile’s `SessionDB`. A canonical-corrupt named profile can therefore be subjected to that live recovery machinery from background work even though this branch’s new regressions prove only correct path selection on healthy databases. `db is not None` plus “the selected path is correct” does not authorize every mutation the selected handle may perform.

Please make the merge topology load-bearing: this needs to depend on #96038 (or carry equivalent fail-closed classification) before it can claim the named-profile path fails closed. Then add a physical cross-profile regression using #86224/#86237’s covering-index fixture:

- root and a second profile remain healthy;
- the target profile’s canonical `messages` payload B-tree is malformed while `COUNT(*)` still succeeds through its covering index;
- an unscoped background write enters through this resolver;
- nothing lands in root or the other profile;
- the target profile gets no FTS rebuild, stale marker, or trigger removal;
- the operation remains visibly failed/queued rather than being retried as an FTS-only event;
- a separate FTS-only target-profile case still self-heals in place.

Without that interlock, the tests prove healthy routing while the actual mutation scope includes structural recovery. The key-owner fix is necessary, but its fail-closed boundary is not true on the current base.

caya8205-2:
New exact head: `373bb3e227b3478d4150738827ad16a8b73e6d9a`, recomposed on `main@4209d371a`.

## P1 — the compression continuation losing its owner

Confirmed and fixed. Your trace was exact: `_append_transcript_message(child_id, msg)` runs before either `_transcript_reroutes` or the `_entries` update is published, so nothing points at the child, the index scan misses, and `_db_for_key(None)` hands back the ambient store — a live handle, so the fail-closed guard never sees it.

I took the second option you offered rather than the first. Carrying the resolved handle as a parameter meant changing `_append_transcript_message`'s signature, which broke four existing suites in `tests/gateway/test_session.py` that stub it with a two-argument callable. Editing those stubs to fit my change seemed like the wrong trade, so ownership is now proven *before* the mutation instead:

- `_session_owner_hints` maps `session_id -> owning routing key` for ids whose owner is proven but not yet published;
- `_owner_key_for_session_id()` consults it after the index scan misses;
- the hint is dropped the moment routing publishes.

Publication ordering is untouched — the queue-migration-before-publication sequence you flagged as load-bearing still runs exactly as it did. No signature changed, and no existing test was modified. The map is read through `getattr` because several suites build stores via `object.__new__` and never run `__init__`.

The regression is physical, as asked: an ended compression parent and a live child existing only in `profiles/fitness/state.db`, no active profile scope, append to the parent, then all four effects asserted — the row lands on the child in the profile store, the pending queue drains, the reroute and the routing entry advance, and root `state.db` is untouched. Without the hint it fails precisely the way you described, on `FOREIGN KEY constraint failed` against root because no child row exists there.

126 passed across `test_multiplex_session_db_profile_scope`, `test_session`, `test_session_store_runtime_stale_guard`, `test_session_store_stale_prune`, `test_multiplex_phase0`, `test_session_store_lock_io`, `test_session_store_expiry_finalized`, `test_compression_concurrent_sessions` — re-run after the rebase, since four of the intervening commits touch `gateway/run.py`.

## On the `_execute_write` recovery interlock

I'd push back on making this PR depend on #96038, and I want to be precise about why rather than just decline.

The misclassification in #97940 is real, and I agree a canonical-corrupt profile subjected to `_try_runtime_fts_rebuild()` / `_enter_fts_fail_open()` is bad. But that machinery is not introduced here, and unscoped background writes were already reaching it before this PR — they were simply reaching it against the **root** store, because that is where ambient resolution sent them. This change moves those writes to the file that actually owns the rows; it does not add a recovery path, change `_execute_write`, or make corruption newly reachable from background work.

The fair part of your point is scope of exposure: writes that previously concentrated on one root database now land across N profile databases. That is a genuine consequence and worth recording. But the alternative to that spread is the data-loss seam this PR exists to close, so I don't think it can be the reason to hold the fix.

Gating a data-loss repair behind another PR that is itself unmerged and unreviewed couples two independent changes and, in practice, stalls both. What I'd propose instead:

- I'll note the interaction explicitly in the PR description and link #97940 and #96038, so the exposure is documented rather than implicit;
- the cross-profile corruption regression you sketched (covering-index fixture, canonical B-tree malformed while `COUNT(*)` still succeeds, target profile gets no rebuild/marker/trigger removal, operation stays visibly failed) belongs naturally with the classification fix — I'm happy to write it against #96038, or as a follow-up here once that lands.

If a maintainer decides the interlock is required before merge, I'll respect that and do the work — but I'd rather that be an explicit maintainer call than something this PR assumes.

## Landing receipt

Still not available to me: the exact-head CI/Docker/Nix runs remain `action_required`, and I can't clear that from here. Everything above is development evidence, as you rightly separate it.

## #66887 scope

Unchanged and explicit: this is not closure for #66887. `_entries` / `_loaded`, `_routing_scope()`, `gateway_routing` and the `sessions.json` mirror remain one process-wide frozen surface that can still disagree with the per-session owner, and that follow-up is ordered after this.

teknium1:
@caya8205-2 — your work is on main, authorship preserved: 5ffaed6e45b (resolve session storage from the key's profile), 58dcc100493 (memoize only resolved homes), 49fb58523f5 (fail closed on a named profile with no home) and a60e04a32ce (prove the compression child's owner before writing). `_named_profile_for_key` / `_db_for_key` / `_owner_key_for_session_id` now live in `gateway/session_persistence.py`, and `tests/gateway/test_multiplex_session_db_profile_scope.py` carries your regressions including the compression-child case.

Thank you for the field evidence (the two sessions with inverted `end_reason`) and for turning each review round into a physical regression. Closing this PR since the commits have landed.

PR #97313, opened 2026-08-28
fix(kanban): require explicit specialist routing

tungtrinh1402:
## Problem

The auto-decomposer prompt presents `null` as the normal fallback for every child assignee. The runtime then normalizes null/blank/unknown assignees to `kanban.default_assignee`.

That collapses specialist routing into the fallback profile. In a live board, implementation, runtime-test, and independent-review leaves all returned `null`; after normalization they were assigned to the same default profile.

## Root cause

`hermes_cli/kanban_decompose.py` asks the LLM for:

```json
"assignee": "<profile name from the roster, or null for default>"
```

and repeatedly tells it to use null when no profile fits. `_normalize_assignee_choice()` correctly prevents unassigned work, but therefore maps that output to the default assignee.

## Fix

Keep runtime fallback behavior for malformed model output, but remove null/default as a suggested normal routing option. Require an exact roster profile name and state the intended specialist ownership examples:

- implementation/debugging -> backend
- test/runtime/soak -> tester
- independent audit/acceptance -> reviewer
- plan/specification only -> planner

## Tests

```text
python3 -m pytest tests/hermes_cli/test_kanban_decompose.py -o 'addopts=' -q
4 passed
```

The added regression test checks that the prompt requires a non-null exact roster assignee and no longer exposes `default_assignee` to the decomposer user message.

## Scope

No dispatcher, DB, or fallback-normalization behavior changed. This narrows only the decomposer instruction that caused valid fallback behavior to be used for ordinary specialist work.

### DIFF
--- a/hermes_cli/kanban_decompose.py
+++ b/hermes_cli/kanban_decompose.py
@@ -58,7 +58,6 @@ matching profile from the available roster.
 You will be given:
   - The original task title and body
   - The list of available profiles (each with name + description)
-  - The fallback "default_assignee" used when no profile fits
 
 Output a single JSON object with this exact shape:
 
@@ -69,7 +68,7 @@ Output a single JSON object with this exact shape:
       {
         "title": "<concrete task title, imperative voice, <= 80 chars>",
         "body":  "<detailed spec for the worker on this child task>",
-        "assignee": "<profile name from the roster, or null for default>",
+        "assignee": "<EXACT profile name from the roster>",
         "parents": [<int>, ...]
       },
       ...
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Prompt-only narrowing that stops the decomposer from collapsing specialist routing into `kanban.default_assignee`. Runtime fallback unchanged — malformed output still safely mapped — but normal instruction now forces exact roster names.

**What it does**
- `hermes_cli/kanban_decompose.py:66` changes `_SYSTEM_PROMPT` assignee line to `"<EXACT profile name from the roster>"` (was `"or null for default"`), adds 4 bullets: every child MUST have non-null exact roster name, examples `implementation→backend / test→tester / audit→reviewer / plan→planner`, complexity not an excuse, verify before return; drops `default_assignee` from `_USER_TEMPLATE` and its interpolation in `decompose_task()`; also updates decomposition-single-work-item prompt.
- Keeps `_normalize_assignee_choice` runtime fallback — only the LLM instruction that caused valid fallback misuse is narrowed.
- Test `tests/hermes_cli/test_kanban_decompose.py:123` pins prompt contains MUST + EXACT + example line, no `"or null for default"` and no `default_assignee` in user template.

**Non-blocking notes**
- Specialist examples are guidance, not enforcement — `verify` bullet helps but model may still map differently if description phrasing differs; keep roster descriptions accurate.
- No dispatcher/DB changes — correct scope; existing validator still rejects invented names.

PR #97317, opened 2026-08-28
feat(simplex): treat an edited inbound message as a correction that supersedes the queued/in-flight original

### LINKED ISSUE #35535: feat(simplex): treat an edited inbound message as a correction that supersedes the in-flight/queued original
## Use case

On SimpleX it's common to realize, a second or two after hitting send, that you made a mistake or left something out — while the agent is already mid-reply, or while the message is still sitting in the processing queue. SimpleX lets you edit a sent message, but today that edit is dropped: the adapter handles `newChatItem(s)` but not `chatItemUpdated`, so the agent acts on the original (mistaken) text.

## Desired behavior

When an inbound SimpleX message is edited (`chatItemUpdated`), treat it as *"I meant this instead,"* correlated to that specific original message:

- **still queued** → replace the queued message's text with the edit;
- **already being processed** → interrupt/steer the in-flight turn with the corrected text (supersede), instead of finishing on the stale original.

## Why this is gateway-core, not adapter-only

An adapter can surface the `chatItemUpdated` event, but correlating an edit to the *specific* in-flight or queued original and superseding it needs gateway-side support: an edit→original message-id correlation, plus a supersede/replace API on the active turn and the pending queue. On its own, an adapter can only re-inject the edited text as a brand-new message (no correlation) — which doesn't achieve "fix the message I just sent."

## Notes

- Adapter prerequisite: handle `chatItemUpdated` and carry the original item id (pairs with populating inbound `message_id` from `itemId`).
- Likely applies to other platforms with inbound edits (Signal receives edits today, but as new messages, not correlated supersedes).

DavidMetcalfe:
## Summary

Implements edited-message supersede for the SimpleX gateway adapter, closing #35535.

Today the SimpleX adapter handles only `newChatItems`; a `chatItemUpdated` event (the terminal-API notification emitted when a user edits a sent message) falls into the unhandled-event log, so the agent acts on the original — mistaken — text. This PR treats an inbound edit as *"I meant this instead"*, correlated to the specific original message by platform message id:

- **Original still queued** (primary pending slot or overflow FIFO) → the queued `MessageEvent`'s text is replaced in place. FIFO order, identity, and metadata are preserved; the correction is silent (no busy-ack).
- **Original in-flight** (the turn's originating message id, now recorded on `TurnState.active_message_id`) → the running agent is redirected with the corrected text, framed as `[User edited their earlier message. Corrected message: "..."]`. Falls back to `steer()` when the agent doesn't support active-turn redirect, and to normal queueing when neither primitive is available. Neither primitive cancels in-flight tools — that is the best available behavior without mutating the transcript (see design notes).
- **Uncorrelated edits** (no queued or in-flight match) → dropped with an `info` log rather than dispatched as a new message. Rationale in *Open questions* below.

## Changes

**`plugins/platforms/simplex/adapter.py`**
- `_handle_chat_item` now extracts `meta.itemId` and sets `message_id=str(itemId)` on every inbound `MessageEvent` (the prerequisite correlation key — previously never populated).
- New `chatItemUpdated` branch routes the event's `AChatItem` through the same parse path with `is_edit=True`, tagging `metadata={"is_edit": True}`. The existing outgoing-direction guard means edits of the bot's own messages are ignored.

**`gateway/session_state.py`**
- `TurnState.active_message_id` records the message id of the event that started the turn; cleared by the existing `TurnState.clear()` at every turn boundary (no new lifecycle to maintain, no leak path).

**`gateway/run.py`**
- `_replace_queued_message()`: first-match, in-place text replacement across both queue levels, modeled on the existing `_clear_goal_pending_continuations` pattern.
- `_handle_edit_supersede()`: the correlation decision — queued replace → in-flight redirect/steer → drop+log. Fully synchronous (no `await` between the queue check, the in-flight check, and the decision), so there is no interleaving window with queue promotion. Inserted in `_handle_message` after authorization, ignored-channel, and plugin-hook guards; adapter producers other than SimpleX get the same behavior free by tagging `metadata["is_edit"]` + `message_id`.

## Design notes

- **Why not rewrite the transcript**: the stale original text stays in history; per the project's prompt-caching constraint, past context is never mutated. The correction reaches the live turn via `redirect`/`steer`, mirroring existing mid-turn steering semantics.
- **Why gateway-core, not adapter-only**: correlating an edit to the specific queued/in-flight original requires the queue/turn structures the adapter can't see — as the issue describes. Any future adapter that tags edits (Signal and Discord both surface them today in other forms) inherits the behavior.
- **Startup-resume path**: the synthetic resume event carries no message id; its `active_message_id` stays `None`, so edits during a resumed turn fall to the uncorrelated policy rather than misfiring.

## Open questions

- **Uncorrelated-edit policy**: dropping (with a log) rather than queueing as a new message is a judgment call — Pro-side reasoning in the design review was that an isolated out-of-context correction ("the") injected into a fresh turn confuses the model; the counter-argument is that a drop silently loses user intent. If maintainers prefer "queue uncorrelated edits as normal messages", it's a two-line change in `_handle_edit_supersede`.

## Tests

- `tests/gateway/test_simplex_plugin.py`: 7 new tests — direct/group edit correlation, outgoing-direction suppression, `message_id` population on the normal path, image-content edits, missing-`itemId` fallback.
- `tests/gateway/test_edit_supersede.py` (new): 15 tests — primary-slot replace, overflow-FIFO replace, in-flight redirect, steer fallback, sentinel-window handling, stale-correlation drop, uncorrelated drop, non-edit pass-through.
- All 38 pass locally; the full `tests/gateway/` suite shows only pre-existing environment flakes present on clean `main`.

Closes #35535

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -9381,6 +9381,14 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                 self._queue_or_replace_pending_event(session_key, event)
                 return True
 
+            elif running_agent is _AGENT_PENDING_SENTINEL:
+                logger.info(
+                    "Edit supersede — edit matched in-flight turn but agent "
+                    "still pending (sentinel); edit dropped for session %s",
+                    session_key,
+                )
+                return True
+
         # 3. Uncoupled edit (#35535 deliberate policy): no queued match,
         #    no in-flight match.  An isolated out-of-context correction
         #    would confuse the model — drop it silently.
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -9262,6 +9262,137 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
             _q_state.conversation.queued_events = kept
         return removed
 
+    def _replace_queued_message(
+        self,
+        session_key: str,
+        adapter: Any,
+        message_id: str,
+        new_text: str,
+    ) -> bool:
+        """Replace the text of the first queued event matching ``message_id``.
+
+        Searches the primary pending slot first, then the overflow FIFO.
+        Mutates the matching event's ``.text`` in place and preserves all
+        other fields and FIFO order.  Returns True if a match was replaced,
+        False if no match was found in either queue level.
+
+        Modeled on ``_clear_goal_pending_continuations`` (same two-level
+        search pattern).
+        """
+        # Primary slot
+        pending_slot = getattr(adapter, "_pending_messages", None) if adapter is not None else None
+        if isinstance(pending_slot, dict):
+            pending_event = pending_slot.get(session_key)
+            if pending_event is not None and getattr(pending_event, "message_id", None) == message_id:
+                pending_event.text = new_text
+                return True
+
+        # Overflow FIFO
+        _q_state = self._peek_session_state(session_key)
+        overflow = _q_state.conversation.queued_events if _q_state else []
+        for ev in overflow:
+            if getattr(ev, "message_id", None) == message_id:
+                ev.text = new_text
+                return True
+
+        return False
+
+    def _handle_edit_supersede(
+        self,
+        event: "MessageEvent",
+        session_key: str,
+        adapter: Any,
+    ) -> bool:
+        """Correlate an inbound edit with queued or in-flight messages.
+
+        Called from the synchronous edit-supersede block in
+        ``_handle_message`` (#35535).  Returns True if the event was handled
+        (queue-replaced, in-flight redirected/steered, or silently dropped);
+        the caller should return ``None`` to skip normal dispatch.  Returns
+        False for non-edit events that should continue through normal
+        processing.
+
+        Design intent (per issue #35535 design review):
+          * Queue match -> replace text in-place, silent (no ack).
+          * In-flight match (active_message_id) -> redirect/steer with
+            framing text.  Neither primitive cancels in-flight tools — this
+            is the best available approach without transcript mutation.
+          * No match (uncorrelated edit) -> silently dropped.  An isolated
+            out-of-context correction would confuse the model.
+
+        SYNCHRONOUS: no await between queued-replace, in-flight check, and
+        decision.  All operations are in-memory attribute writes.
+        """
+        _is_edit = bool(event.metadata.get("is_edit", False) if event.metadata else False)
+        _msg_id = event.message_id
+        if not _is_edit or not _msg_id:
+            return False  # not an edit — continue normal dispatch
+
+        # 1. Try queued supersede
+        if self._replace_queued_message(session_key, adapter, _msg_id, event.text or ""):
+            logger.info(
+                "Edit supersede — replaced queued message_id=%s for session %s",
+                _msg_id, session_key,
+            )
+            return True
+
+        # 2. Try in-flight redirect / steer
+        _q_state = self._peek_session_state(session_key)
+        if _q_state and _q_state.turn.active_message_id == _msg_id:
+            running_agent = _q_state.turn.agent
+            if running_agent is not None and running_agent is not _AGENT_PENDING_SENTINEL:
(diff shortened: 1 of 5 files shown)


Enough1122:

**Overall:** Thoughtful edit-supersede implementation that treats a `chatItemUpdated` edit as a correction correlated by `itemId`/`message_id`. Queued text is replaced in-place (order preserved), in-flight turns are `redirect`/`steer`-framed, and uncorrelated edits are dropped — all synchronous on the gateway's single event loop with no await gaps (so no lock needed).

**What it does**
- `plugins/platforms/simplex/adapter.py:173` adds `chatItemUpdated` branch (stable `itemId → message_id=str(itemId)`, `metadata={"is_edit":True}`), guards outgoing edits via existing direction check, and tags new `newChatItems` similarly; file-caption-only edit intentional.
- `gateway/session_state.py:73` adds `TurnState.active_message_id` (cleared with `clear()`).
- `gateway/run.py:9262` adds `_replace_queued_message(session_key, adapter, message_id, new_text)` (primary `_pending_messages[session_key]` then `overflow FIFO` `queued_events`, mutates `.text` in place, preserves FIFO) and `_handle_edit_supersede(event, session_key, adapter)` (edit+id guard, queued replace silent, in-flight `active_message_id` match → `redirect(framed)` → `steer` fallback, both best-effort, no tool cancellation, uncorrelated → info log + drop; explicitly notes single-loop synchrony avoids interleaving).
- Tests `tests/gateway/test_edit_supersede.py:234` and `tests/gateway/test_simplex_plugin.py:411` cover queued primary/overflow replacement, no-match, in-flight redirect/steer, dispatcher wiring, and simplex direct/group/outgoing/message_id propagation.

**Non-blocking notes**
- `redirect` vs `steer` is best-available: neither cancels in-flight tools — documented as without transcript mutation; mid-tool correction will interleave — acceptable tradeoff.
- `uncorrelated edit → dropped` is deliberate to avoid out-of-context model confusion — ensure operator expectation documented (edit of already-consumed turn is not replayed).
- Session-scoped correlation prevents cross-chat overwrite in groups (key derived from `platform+chat_id`) — correct.

DavidMetcalfe:
@Enough1122 Thanks for the review.

One correction on the summary: the `newChatItems` path is not tagged as an edit — only the `chatItemUpdated` branch sets `metadata={"is_edit": True}` (plugins/platforms/simplex/adapter.py:676-677, gated on the `is_edit` parameter that branch passes); normal inbound messages gain only `message_id=str(itemId)`. The gateway's supersede logic keys off `is_edit` + `message_id`, so ordinary messages are unaffected.

(Note: the review's line references had drifted — the `chatItemUpdated` branch is at adapter.py:453, `TurnState.active_message_id` at session_state.py:78, `_replace_queued_message` at run.py:9265.)

On the three non-blocking notes — all as designed: the redirect/steer no-tool-cancellation tradeoff is documented in the PR body's design notes, the uncorrelated-edit drop policy is under "Open questions" in the PR body awaiting maintainer preference, and group sessions are keyed per-chat so cross-user supersede can't happen.

Enough1122:
Thanks for the correction — you're right. Re-reading at head `b300eda6e9`, only the `chatItemUpdated` branch sets `metadata={"is_edit": True}` (adapter.py:676-677, gated on `is_edit`), while `newChatItems` just gains `message_id=str(itemId)` — ordinary inbound messages are unaffected and the gateway's `is_edit + message_id` correlation is clean. My summary line conflated the two paths; noted and corrected.

The per-chat session keying and the documented redirect/steer and uncorrelated-drop tradeoffs are as you say — no action needed from me.

DavidMetcalfe:
@Enough1122 Appreciate the follow-up — glad the distinction is clear now. No further action needed on this thread.

Enough1122:
Ack — correction noted and accepted. Thanks for clarifying the `is_edit` gating (only `chatItemUpdated` branch tags `metadata={"is_edit": True}`) and the documented tradeoffs. No further items from me.

Issue #88141, opened 2026-08-17
[Feature]: Native video fast-path for video_analyze — route video into the main model's multimodal tool-result channel when the model supports video input


uraniumchonk:
## Problem or Use Case

Image analysis already has a native fast-path: when the active main model is vision-capable and its provider accepts media inside tool results, `vision_analyze` short-circuits the auxiliary describer and returns the raw pixels as a multimodal tool-result envelope (`_should_use_native_vision_fast_path()` / `_build_native_vision_tool_result()` in `tools/vision_tools.py`). The main model then sees the image on its next turn — no auxiliary call, no text-summary information loss.

`video_analyze` has no equivalent. `video_analyze_tool()` always sends the full video as a base64 `video_url` block through the auxiliary client (`async_call_llm(task="vision", ...)`) and returns a text JSON `{success, analysis}` — even when the active main model natively supports video input (e.g. Gemini 3, MiniMax-M3). Consequences:

- Temporal/action/audio detail is reduced to a text summary; the main model's native video understanding is never used.
- Every video analysis pays a full auxiliary-model round trip even when the main model could process it directly.
- Video attachments are still handed to the agent as bare disk paths (see #49565), forcing the same auxiliary path.

## Proposed Solution

Mirror the image native fast-path for video:

1. `_should_use_native_video_fast_path()` — same shape as the image check: main model input modalities include `video` AND the provider's tool-result channel can carry video media parts.
2. `_build_native_video_tool_result()` — wrap the existing base64 `video_url` data URL into a multimodal tool-result envelope; the agent loop unwraps it into a `tool`-role content list and provider adapters translate it (OpenAI-style media parts, Gemini `inlineData`/`fileData`, MiniMax `video` blocks).
3. Keep the auxiliary path as the fallback whenever the main model or provider cannot accept video in tool results. Anthropic today only accepts image blocks in `tool_result` (see the #72371 discussion), so this fallback stays essential. Optionally gate with a config key (e.g. `video_input_mode: native|auxiliary`) so users can force either direction.

Prior art already in flight: Gemini adapter `video_url`/inline-video translation (#81304, #65043), MiniMax Messages video input (#84524), frame-extraction fallback for local VLM endpoints (#82996), and a separate `auxiliary.video` config block (#27313 / PR #31107).

## Alternatives Considered

[...]

PR #72516, opened 2026-07-27
fix(vision): fallback to image_url when provider rejects video_url (#72275)

JonthanaHanh:
## Summary
Fixes #72275.

Most OpenAI-compatible providers (OpenRouter, Ollama Cloud, etc.) accept `image_url` content blocks but reject `video_url` blocks with a 400 error. `video_analyze_tool` sent video exclusively as `video_url` with no fallback.

## Changes
- `tools/vision_tools.py`: Wrap the initial `async_call_llm` call in a try/except. When the error matches video-format hints, retry with the same base64 data sent as an `image_url` content block. Update error message to reflect both formats were attempted.

## Testing
- `py_compile`: OK
- `ruff check`: All checks passed
- The fallback only triggers on video-format-related errors; other errors propagate normally.

PR #97318, opened 2026-08-28
fix(vision): negotiate native video dialects and fall back to frames

### LINKED ISSUE #72275: [Bug]: video_analyze sends raw video base64 that most providers reject with 400
## Description

`video_analyze` tool base64-encodes the entire video file into a `data:` URL and sends it to the configured auxiliary vision provider inside an OpenAI-style `video_url` content block. Most providers that are OpenAI-compatible for _images_ do not accept video input via this message shape, returning a `400 invalid message format` error. The error handler then misleadingly reports "the model does not support video" regardless of the actual reason.

## Root Cause

In `tools/vision_tools.py` (lines 1721-1726), the video payload is hard-coded as:

```python
{
    "type": "video_url",
    "video_url": {
        "url": video_data_url,
    },
}
```

This is sent to the auxiliary vision provider (configured under `auxiliary.vision`). While PR #38999 fixed the wire format for Anthropic-protocol endpoints (MiniMax/Bedrock), the tool still fails for OpenAI-compatible providers like Ollama Cloud, OpenRouter, and most others, because:

1. Most OpenAI-compatible vision APIs accept `image_url` blocks but not `video_url` blocks
2. There is no fallback mechanism: if the provider rejects video input, the tool errors out rather than trying frame extraction
3. The error message is generic and misleading

## Steps to Reproduce

1. Configure auxiliary vision model that supports images but not native video input:
   ```yaml
   auxiliary:
     vision:
       provider: ollama-cloud
       model: gemma4:31b
   ```
2. Call `video_analyze` with any local or remote video:
   ```
   video_analyze(video_url="/tmp/test.mp4", question="What happens?")
   ```
3. Observe: `400 invalid message format` error

## Expected Behavior

The `video_analyze` tool should either:
- Detect that the provider does not support video input and automatically fall back to frame extraction (ffmpeg → individual frames → `vision_analyze` each frame)
- Or at minimum provide a clear error message explaining that the configured provider doesn't accept video and suggesting the ffmpeg-frames workaround

## Suggested Fix

In `tools/vision_tools.py`, before sending the video to the LLM, check if the provider supports video. If not, fall back to frame extraction:

```python
# Pseudo-code for fallback
import subprocess, tempfile
if not provider_supports_video(resolved_provider):
    frames = extract_frames_ffmpeg(video_path, interval_sec=10)
    descriptions = []
    for frame in frames:
        desc = analyze_image(frame, question)  # reuse vision_analyze logic
        descriptions.append(desc)
    return synthesize(descriptions, question)
```

## Environment

- Hermes Agent: 2026.7.20 (Homebrew)
- Python: 3.14.6
- OS: macOS 26.5.2
- Auxiliary vision: ollama-cloud / gemma4:31b

## Related

- #37219 (fixed by #38999) — fixed Anthropic wire format but not the broader compatibility issue
- #70948 — related fallback issue for `vision_analyze` (images)

andrexibiza:
## Current landing gate — 2026-08-30

- **Base/current upstream main at rematerialization:** `2a598aad1c398e95b3325a0f100f5c28efa63d12`
- **Current upstream main cross-check:** `2a598aad1c398e95b3325a0f100f5c28efa63d12`
- **Exact head:** `1e53d8d75e7b6cb82dbf8bb1812b5e04a61c27dc`
- **Tree:** `f8c0d12151946387a325c81179292a19ef872845`
- **Submitted train:** exactly **1 commit / 6 files**, open, non-draft, mergeable
- **Parent:** exact current main
- **Every modified file:** below the repository 2,000-line ceiling
- **Hosted acceptance:** **GREEN on the exact submitted object** — CI `33307205290`, Docker `33307204886`, and Nix `33307204902` all completed successfully. CI job `99246181899` (`All required checks pass`) is successful.

The previous red head exposed a regression-harness boundary bug, not a remaining product-buffering bug: product code already rejects from `Path.stat().st_size` before `read_bytes()`, but the regression created a file one byte above the conservative universal raw-byte budget rather than one byte above the actual MIME-specific `data:video/mp4;base64,` budget. The submitted object fixes the witness to use `_max_raw_video_bytes("video/mp4") + 1`, so the test now targets the real pre-buffer rejection boundary.

A temporary replacement PR (#98502) was opened solely to test whether a fresh PR-open event would bootstrap hosted workflows for this exact commit. It produced no run and was immediately closed unmerged to preserve this PR as the single active carrier. Hosted execution later arrived on #97318 itself; only those exact-head runs are used as acceptance evidence.

## What does this PR do?

`video_analyze` currently sends one hard-coded `video_url` block containing a full `data:video/...;base64,...` payload. That shape works only for a subset of providers: image-compatible OpenAI endpoints reject it, while llama.cpp accepts native video under a different typed-content dialect.

This PR replaces the false binary of “native `video_url` or generic model-capability error” with a bounded, response-driven transport ladder:

1. Preserve the existing `video_url` request as the first attempt.
2. When the provider rejects that content dialect, retry llama.cpp’s documented `input_video` form with raw base64 bytes.
3. When native video is rejected but image input is available, extract bounded chronological JPEG frames with `ffmpeg`/`ffprobe` and submit real `image_url` parts.
4. Stop immediately for non-negotiable failures such as authentication, billing, rate limits, timeouts, payload limits, server errors, and missing models.

The provider response authorizes each transition. Endpoint location and guessed model names do not.

The implementation moves runtime ownership into the bounded `tools/video_analysis/` package while preserving the legacy `tools.vision_tools` public/patch surface and deterministic registry discovery.

### Review repairs included in the exact submitted object

- one canonical frame-extraction owner in `tools/video_analysis/frames.py`;
- leading-status-only fallback parsing rather than matching arbitrary embedded numbers;
- raw-file size rejection before `Path.read_bytes()` using the exact MIME-specific data-URL budget;
- configured video timeout values are honored rather than silently clamped;
- deterministic runtime ownership through the bounded package/discovery bridge.

### Sources and prior work

- #72275 by Catmittee — reporter and reproduction.
- #72516 by JonthanaHanh — established the rejection/retry seam, but its `image_url` retry still carried video bytes.
- #82996 by x7peeps — real frame-extraction direction.
- Current-main path/error-classification qualification from the #72275 discussion.
- llama.cpp server source/documentation establishes `input_video` as a native-video request dialect.

This PR preserves those contributions while remaining the single active settlement owner for this overlapping transport/fallback slice.

## Related Issue

Fixes #72275.

Related: #88141.

## Type of Change

- [ ] ✨ New feature
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests
- [x] ♻️ Refactor
- [ ] 🎯 New skill

## Changes Made

- `tools/video_analysis/core.py`
  - response-driven `video_url → input_video → image_frames` negotiation;
  - MIME-specific encoded-payload budget checked from file metadata before buffering;
  - fatal vs negotiable provider-error classification;
  - configured timeout/temperature resolution without silent timeout clamping;
  - structured settled-transport/frame-count receipts.
- `tools/video_analysis/frames.py`
  - one bounded frame-extraction owner;
  - chronological midpoint sampling;
  - argv-only `ffmpeg`/`ffprobe`, disabled stdin, explicit timeouts, one ffmpeg thread;
  - bounded dimensions/encoded payload and JPEG magic validation.
- `tools/video_analysis/__init__.py`
  - canonical bounded package owner with legacy compatibility exports.
- `tools/vision_video_analysis.py`
  - deterministic registry/discovery settlement after legacy vision-tool discovery.
- `tests/tools/test_video_transport_negotiation.py`
  - native dialect retry, image-frame fallback, fatal-error refusal, extraction/runtime compatibility, real ffmpeg path, and registry settlement.
- `tests/tools/test_video_transport_regressions.py`
  - single extraction ownership, leading-only status fallback, pre-buffer oversize refusal at the actual `video/mp4` budget, and configured-timeout preservation.

## How to Test

```bash
python -m pytest \
  tests/tools/test_video_analyze.py \
  tests/tools/test_video_transport_negotiation.py \
  tests/tools/test_video_transport_regressions.py \
  tests/tools/test_vision_tools.py \
  tests/tools/test_computer_use_vision_routing.py \
  tests/tools/test_vision_region.py \
  tests/tools/test_vision_native_fast_path.py \
  -q

python -m ruff check \
  tools/video_analysis \
  tools/vision_video_analysis.py \
  tests/tools/test_video_transport_negotiation.py \
  tests/tools/test_video_transport_regressions.py

python scripts/check-windows-footguns.py \
  tools/video_analysis \
  tools/vision_video_analysis.py \
  tests/tools/test_video_transport_negotiation.py \
  tests/tools/test_video_transport_regressions.py

git diff --check 2a598aad1c398e95b3325a0f100f5c28efa63d12..1e53d8d75e7b6cb82dbf8bb1812b5e04a61c27dc
```

### Adversarial acceptance cases

- llama.cpp-style rejection of `video_url` advances to `input_video` with raw base64;
- image-only fallback sends actual JPEG frames, never video bytes mislabeled as images;
- auth/rate-limit/server/payload failures do not get reclassified as dialect negotiation;
- a `video/mp4` one byte above its MIME-specific raw-byte budget is rejected from `stat().st_size` before `read_bytes()` executes;
- embedded strings such as `429 bytes` are not mistaken for HTTP status 429;
- configured timeout `60` remains `60.0`;
- one canonical frame-extraction implementation owns probing, timestamps, and JPEG conversion.

### Code

- [x] I've read the Contributing Guide.
- [x] The single submitted commit follows Conventional Commits and includes DCO sign-off.
- [x] Existing work and contributor provenance are reconciled.
- [x] This PR contains only the video transport/fallback repair and its regressions.
- [x] Tests cover the original defect and review-discovered sibling classes.
- [x] Hosted exact-head CI/Docker/Nix acceptance — CI `33307205290`, Docker `33307204886`, and Nix `33307204902` all succeeded on `1e53d8d75e7b6cb82dbf8bb1812b5e04a61c27dc`; CI aggregate `All required checks pass` job `99246181899` succeeded.

- [x] No public config key changed.
- [x] No contributor workflow changed.
- [x] No model-facing tool schema change requires separate documentation.
- [x] Every submitted file is below 2,000 lines.
- [x] Submitted train is one commit directly on current main.

## Screenshots / Logs

No UI surface changes.

Historical runs belong to historical heads. Exact head `1e53d8d75e7b6cb82dbf8bb1812b5e04a61c27dc` is fully green across CI, Docker, and Nix.

### DIFF
--- a/tests/tools/test_video_transport_negotiation.py
+++ b/tests/tools/test_video_transport_negotiation.py
@@ -1,0 +1,407 @@
+"""Regression coverage for provider video-transport negotiation."""
+
+from __future__ import annotations
+
+import asyncio
+import base64
+import json
+import shutil
+import subprocess
+from io import BytesIO
+from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from tools import video_analysis
+from tools.video_analysis import (
+    VideoFrame,
+    VideoFrameExtractionError,
+    video_analyze_tool,
+)
+from tools.video_analysis.frames import extract_video_frames
+
+
+class ProviderError(RuntimeError):
+    def __init__(self, status_code: int, message: str):
+        super().__init__(message)
+        self.status_code = status_code
+
+
+def _response(text: str = "ok") -> MagicMock:
+    response = MagicMock()
+    response.choices = [MagicMock()]
+    response.choices[0].message.content = text
+    return response
+
+
+def _run(coro):
+    return asyncio.run(coro)
+
+
+def test_llama_cpp_rejection_retries_documented_input_video(tmp_path):
+    video = tmp_path / "clip.mp4"
+    video_bytes = b"native-video-bytes"
+    video.write_bytes(video_bytes)
+    calls = []
+
+    async def call_llm(**kwargs):
+        calls.append(kwargs)
+        if len(calls) == 1:
+            raise ProviderError(
+                400,
+                "HTTP 400: unsupported content[].type: video_url",
+            )
+        return _response("native llama.cpp video")
+
+    with (
+        patch("tools.vision_tools.async_call_llm", side_effect=call_llm),
+        patch(
+            "tools.vision_tools.extract_content_or_reasoning",
+            return_value="native llama.cpp video",
+        ),
+        patch(
+            "tools.video_analysis.core._extract_video_frames_async",
+            new_callable=AsyncMock,
+        ) as extract_frames,
+    ):
+        result = _run(
+            video_analyze_tool(str(video), "Describe the timeline")
+        )
+
+    data = json.loads(result)
+    assert data["success"] is True
+    assert data["transport"] == "input_video"
+    assert len(calls) == 2
+    first_content = calls[0]["messages"][0]["content"]
+    second_content = calls[1]["messages"][0]["content"]
+    assert first_content[1]["type"] == "video_url"
+    assert second_content[1]["type"] == "input_video"
+    raw_payload = second_content[1]["input_video"]["data"]
+    assert not raw_payload.startswith("data:")
+    assert base64.b64decode(raw_payload) == video_bytes
+    extract_frames.assert_not_awaited()
+
+
+def test_bare_400_negotiates_input_video_before_frames(tmp_path):
+    video = tmp_path / "clip.mp4"
+    video.write_bytes(b"video")
+    calls = []
+
+    async def call_llm(**kwargs):
+        calls.append(kwargs)
+        if len(calls) == 1:
+            raise ProviderError(400, "Bad Request")
+        return _response("accepted")
+
+    with (
+        patch("tools.vision_tools.async_call_llm", side_effect=call_llm),
+        patch(
+            "tools.vision_tools.extract_content_or_reasoning",
+            return_value="accepted",
+        ),
+        patch(
+            "tools.video_analysis.core._extract_video_frames_async",
+            new_callable=AsyncMock,
+        ) as extract_frames,
+    ):
+        result = _run(video_analyze_tool(str(video), "Describe this"))
+
+    data = json.loads(result)
+    assert data["success"] is True
+    assert data["transport"] == "input_video"
+    assert [
+        call["messages"][0]["content"][1]["type"] for call in calls
+    ] == ["video_url", "input_video"]
+    extract_frames.assert_not_awaited()
+
+
+def test_explicit_image_only_rejection_skips_input_video(tmp_path):
+    video = tmp_path / "clip.mp4"
+    video.write_bytes(b"video")
+    frames = [VideoFrame(0.5, "data:image/jpeg;base64,RlJBTUU=")]
+    calls = []
+
+    async def call_llm(**kwargs):
+        calls.append(kwargs)
+        if len(calls) == 1:
+            raise ProviderError(400, "model does not support video input")
+        return _response("visible frame")
+
+    with (
+        patch("tools.vision_tools.async_call_llm", side_effect=call_llm),
+        patch(
+            "tools.vision_tools.extract_content_or_reasoning",
+            return_value="visible frame",
+        ),
+        patch(
+            "tools.video_analysis.core._extract_video_frames_async",
+            new_callable=AsyncMock,
+            return_value=frames,
+        ) as extract_frames,
+    ):
+        result = _run(video_analyze_tool(str(video), "Describe this"))
+
+    data = json.loads(result)
+    assert data["success"] is True
+    assert data["transport"] == "image_frames"
+    assert len(calls) == 2
+    assert calls[0]["messages"][0]["content"][1]["type"] == "video_url"
+    assert [
+        part["type"] for part in calls[1]["messages"][0]["content"]
+    ] == ["text", "image_url"]
+    extract_frames.assert_awaited_once_with(video)
+
+
+def test_image_fallback_sends_real_ordered_jpeg_frames(tmp_path):
+    video = tmp_path / "clip.mp4"
+    video.write_bytes(b"video")
+    calls = []
+    frames = [
+        VideoFrame(0.5, "data:image/jpeg;base64,RlJBTUUx"),
+        VideoFrame(2.5, "data:image/jpeg;base64,RlJBTUUy"),
+    ]
+
+    async def call_llm(**kwargs):
+        calls.append(kwargs)
+        if len(calls) == 1:
+            raise ProviderError(400, "unsupported content[].type video_url")
+        if len(calls) == 2:
+            raise ProviderError(400, "unsupported content[].type input_video")
+        return _response("frame analysis")
+
+    with (
+        patch("tools.vision_tools.async_call_llm", side_effect=call_llm),
+        patch(
+            "tools.vision_tools.extract_content_or_reasoning",
+            return_value="frame analysis",
+        ),
+        patch(
+            "tools.video_analysis.core._extract_video_frames_async",
+            new_callable=AsyncMock,
+            return_value=frames,
+        ) as extract_frames,
+    ):
+        result = _run(
+            video_analyze_tool(str(video), "Describe the timeline")
+        )
+
+    data = json.loads(result)
+    assert data["success"] is True
+    assert data["transport"] == "image_frames"
+    assert data["frame_count"] == 2
+    extract_frames.assert_awaited_once_with(video)
+
+    frame_content = calls[2]["messages"][0]["content"]
+    assert [part["type"] for part in frame_content] == [
+        "text",
+        "image_url",
+        "image_url",
+    ]
+    assert [
+        part["image_url"]["url"] for part in frame_content[1:]
+    ] == [frame.data_url for frame in frames]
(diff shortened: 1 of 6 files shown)


andrexibiza:
Exact-head verification receipt for `3c4c788fa8a491ef9d7c6291fdcf72cf9d189374` against base pin `baa344dee76993f0444c18fc59a69738ccb339d0`:

- [CI](https://github.com/NousResearch/hermes-agent/actions/runs/33193876568): success — Python 3.11/3.12 tests, Windows smoke tests, Ruff, Windows-footgun guard, attribution, documentation, packaging/install, compatibility, dependency, and security gates.
- [Nix flake check](https://github.com/NousResearch/hermes-agent/actions/runs/33193875559): success.
- [Docker build, test, and publish](https://github.com/NousResearch/hermes-agent/actions/runs/33193875469): success.
- Non-executable label-rerun housekeeping concluded skipped/cancelled under concurrency; it did not replace or invalidate the successful code/build workflows above.
- Topology: one commit, five changed paths, zero commits behind the pinned base.
- GitHub merge state: clean.

Enough1122:

The negotiation ladder is well designed: try the OpenAI-style `video_url`, retry the llama.cpp `input_video` dialect only on a wire-format rejection, then fall back to bounded chronological JPEG frames — and every transition is authorized by the provider's actual response, never by guessing the endpoint or model name. The bounded ffmpeg subprocess calls (argv vectors, stdin disabled, per-call timeout, private temp dir with Windows-safe cleanup) and the evidence-note prompt that stops the model from inventing audio/continuity are both the right calls. The regression tests are genuinely thorough, including the real-ffmpeg integration test.

Points to consider:

1. **The frame-extraction implementation is duplicated and has already diverged.** `core.py` ships a full `_extract_video_frames`/`_probe_video_duration`/`_frame_timestamps`/`_jpeg_data_url` (~`core.py` L793-924), and `frames.py` ships near-identical copies of the same four functions (~`frames.py` L1581-1774). Then `tools/video_analysis/__init__.py` (L431-432) silently overwrites core's globals with frames' versions at import, so core's copy is dead code that only exists to be replaced. Worse, the two copies have already drifted: core's ffmpeg filter is `scale=w='min({max_side},iw)':h=-2` (constrains only width), while frames' is `scale={max_side}:{max_side}:force_original_aspect_ratio=decrease:force_divisible_by=2` (the bounded variant the tests actually assert). Per the repo's extend-don't-duplicate rule, keep one implementation (frames.py is the better one — the bounded aspect handling) and have core import it, so the two can't silently disagree again.

2. **`_status_code` regex over-matches any 3-digit number** (`core.py` L1004-1012): `re.search(r"\b(?:http\s*)?(\d{3})\b", str(exc))` will read *any* three consecutive digits in an error message as an HTTP status. Provider messages routinely embed such numbers for unrelated reasons — e.g. `"stream ended after 429 bytes"`, `"model returned 404 tokens"`, `"attempt 429: retrying"` — and each would be misclassified as a fatal status, skipping the whole negotiation ladder (`_is_fatal_request_error` returns True → `_should_try_input_video`/`_should_try_frames` return False). Prefer `exc.status_code` / `response.status_code` only (already tried first), and if the string fallback must stay, anchor it to the leading position (`^(?:HTTP )?(\d{3})\b`) or restrict it to the first 20 chars.

3. **The 50 MB cap is enforced after reading the whole file into memory** (`core.py` L1323-1334): `_video_to_base64_data_url` calls `video_path.read_bytes()`, and only then is the encoded length checked against `_MAX_VIDEO_BASE64_BYTES`. `video_size_bytes` is already available at that point (`stat().st_size` at L1308) — a pre-check of `st_size` against the base64-inflated limit (~37 MB of raw video) would reject oversized inputs without buffering 50 MB first.

4. **`_resolve_video_call_settings` clamps the timeout to ≥ 180 s** (`core.py` L1212-1231): `timeout = max(float(configured_timeout), 180.0)`. That's defensible for a slow video round-trip, but it means a user who sets `auxiliary.video.timeout: 60` to fail fast is silently overridden. If the floor is intentional, a one-line comment saying so would prevent a confused config bug report; otherwise use the configured value directly.

5. **Registry dance depends on import order** (`core.py` L1545 + `__init__.py` L489 + `tools/vision_video_analysis.py` L1801): three registrations of `video_analyze` with `override=True`, relying on sorted module discovery to settle on the package owner. The discovery-bridge test pins the final state, which is good — just worth a comment near the `override=True` sites documenting that the ordering is load-bearing, so a future rename (e.g. `vision_video_analysis.py` sorting later than expected) doesn't silently regress which handler wins.

andrexibiza:
Exact-object closure cross-check is complete on `1e53d8d75e7b6cb82dbf8bb1812b5e04a61c27dc` / tree `f8c0d12151946387a325c81179292a19ef872845`.

- current upstream `main` is still the exact parent `2a598aad1c398e95b3325a0f100f5c28efa63d12`;
- submitted train is exactly 1 surviving commit / 6 files;
- CI `33307205290` — success, including `All required checks pass` job `99246181899`;
- Docker `33307204886` — success;
- Nix `33307204902` — success;
- PR remains open, non-draft, and mergeable.

The prior missing-hosted-execution boundary is therefore closed on the actual submitted object. The PR body has been refreshed so no historical/pending CI claim remains. #97318 is substantively ready to land; #72275 should remain open until the PR is actually merged.

PR #97319, opened 2026-08-28
fix(desktop): exempt the main alias from the stale-aux banner

### LINKED ISSUE #97310: [Bug]: Desktop stale-aux banner misfires when auxiliary provider is the 'main' alias
## Bug Description

In the desktop app's Settings → Model page, the "stale auxiliary task" banner misfires when an auxiliary task's provider is set to the `main` alias:

> ⚠ 1 auxiliary task (vision) still run on `main`, not your main model.

`main` is a valid backend alias meaning "follow the current main provider". For example, Moonshot's official Hermes integration guide (kimi.com/code/docs/third-party-tools/hermes.html) instructs users to configure:

```yaml
auxiliary:
  vision:
    provider: main
    model: kimi-k3
```

With this configuration the vision task correctly runs on the main model, yet the banner claims it does not — confusing users into thinking their aux task is pinned to a dead or foreign provider.

## Steps to Reproduce

1. Use any provider as the main model (e.g. `kimi-coding` with model `kimi-k3`).
2. Set in `config.yaml`:
   ```yaml
   auxiliary:
     vision:
       provider: main
       model: kimi-k3
   ```
3. Open the desktop app → Settings → Model.
4. The amber banner appears: "1 auxiliary task (vision) still run on `main`, not your main model."

## Expected Behavior

No warning: `provider: main` means the task follows the main model, which is exactly the state the banner wants users to be in.

## Actual Behavior

The warning fires on a correct, working configuration.

## Root Cause

`apps/desktop/src/app/settings/model-settings.tsx`, `persistentStaleAux` (~line 493-507):

```ts
return auxiliary.tasks
  .filter(entry => {
    const p = (entry.provider ?? '').toLowerCase()
    return p && p !== 'auto' && p !== mainProvider
  })
```

The stale check exempts `''` and `'auto'` but not `'main'`. The backend resolves the `main` alias correctly (`agent/auxiliary_client.py`, `resolve_vision_provider_client`, including the active-provider fallback at ~line 7441-7450), and image input works end-to-end — so this is a GUI-only false positive.

## Suggested Fix

Exempt the `main` alias in `persistentStaleAux`:

```ts
return p && p !== 'auto' && p !== 'main' && p !== mainProvider
```

(Or resolve `main` to the active main provider slug before comparing.)

## Environment

- Hermes Agent: v0.20.6 (2026.8.27)
- OS: Windows 11
- Surface: Electron desktop app, Settings → Model
- Main provider: kimi-coding (`https://api.kimi.com/coding/v1`), model `kimi-k3`

Jackal991:
## Summary

The desktop app's Settings → Model page shows a false "stale auxiliary task" warning when an auxiliary task's provider is set to the `main` alias. `main` is a valid backend alias meaning "follow the current main provider" (e.g. Moonshot's official Hermes integration guide configures `auxiliary.vision.provider: main`). The banner claimed the task "still run[s] on `main`, not your main model" even though `main` correctly follows the main model — a GUI-only false positive.

## Changes

- `apps/desktop/src/app/settings/model-settings.tsx`: exempt the `main` alias in the `persistentStaleAux` filter, alongside the existing `auto` and current-main-provider exemptions.
- `apps/desktop/src/app/settings/model-settings.test.tsx`: add a regression test asserting an aux slot pinned to `provider: 'main'` does not trip the stale-aux banner.

## Tests

- New regression test fails on the buggy code (banner renders "1 auxiliary task (Vision) still run on `main`") and passes with the fix.
- `model-settings.test.tsx`: 23/23 pass. Full settings dir: 303 tests pass, no regressions.

Closes #97310

### DIFF
--- a/apps/desktop/src/app/settings/model-settings.test.tsx
+++ b/apps/desktop/src/app/settings/model-settings.test.tsx
@@ -409,6 +409,27 @@ describe('ModelSettings', () => {
     // Banner present on load, no switch required.
     expect(await screen.findByText(/still run on/)).toBeTruthy()
   })
+
+  it('does not flag an aux slot pinned to the main alias as stale', async () => {
+    // `main` is a valid backend alias meaning "follow the current main
+    // provider" (e.g. Moonshot's guide configures auxiliary.vision.provider:
+    // main). It resolves to the main model, so it must not trip the stale-aux
+    // banner. #97310
+    getAuxiliaryModels.mockResolvedValueOnce({
+      main: { provider: 'nous', model: 'hermes-4' },
+      tasks: [{ task: 'vision', provider: 'main', model: 'kimi-k3', base_url: '' }]
+    })
+
+    await renderModelSettings()
+
+    // Wait for the aux data to load and the task rows to render before
+    // asserting the banner is absent — otherwise the assertion runs before
+    // the async load resolves and passes vacuously.
+    await screen.findAllByRole('button', { name: 'Set to main' })
+
+    // No banner on load — `main` correctly follows the main model.
+    expect(screen.queryByText(/still run on/)).toBeNull()
+  })
 })
 
 describe('ModelSettings MoA preset editor', () => {
--- a/apps/desktop/src/app/settings/model-settings.tsx
+++ b/apps/desktop/src/app/settings/model-settings.tsx
@@ -501,7 +501,7 @@ export function ModelSettings({ onMainModelChanged, scopeProfile }: ModelSetting
       .filter(entry => {
         const p = (entry.provider ?? '').toLowerCase()
 
-        return p && p !== 'auto' && p !== mainProvider
+        return p && p !== 'auto' && p !== 'main' && p !== mainProvider
       })
       .map(entry => ({ task: entry.task, provider: entry.provider, model: entry.model }))
   }, [auxiliary, mainModel])


Enough1122:

**Overall:** One-line rendering fix that stops a false stale-aux banner when an aux task is pinned to `main`. `main` means "follow the current main provider" — correctly follows main model, not a stale pin.

**What it does**
- `apps/desktop/src/app/settings/model-settings.tsx:501` extends `persistentStaleAux` filter from `p !== 'auto' && p !== mainProvider` to also `p !== 'main'` (case-folded), alongside existing `auto` exemption.
- Test `tests/gateway/test?...` actually `apps/desktop/src/app/settings/model-settings.test.tsx:409` adds `does not flag aux slot pinned to main alias as stale` (mock `main:{provider:nous,model:hermes-4}`, `tasks:[{task:vision,provider:main,...}]` → no banner after async load; await `Set to main` buttons before asserting absent to avoid vacuous pass).

**Non-blocking notes**
- Matches Moonshot integration guide example (`auxiliary.vision.provider: main`) — good operator-facing fix.
- No backend logic touched — banner is GUI-only false positive; correct scope.

Jackal991:
Agreed on the overall — this is a one-line GUI-only fix, and the `main` alias correctly means "follow the current main provider," so exempting it from the stale-aux banner is the right call. The test change (awaiting the `Set to main` buttons before asserting the banner is absent) is exactly the right guard against a vacuous pass — good catch on that.

No backend logic touched, correct scope. Nothing to change here.

teknium1:
Fixed on `main` by #110208, which salvaged the earlier #97763 by @tutan0558 (same fix, plus the backend `_stale_aux_pins` half) with authorship preserved — `main` is now exempt in `staleAuxAssignments` and in the switch report. Thanks @Jackal991 for the independent fix; closing as landed.

PR #96329, opened 2026-08-27
fix(desktop): a project-tree root load only satisfies its own connection (follow-up to #90334)

vanhoof:
## What does this PR do?

**Stacked on #90334** — the first two commits are that PR's, unchanged; only the last commit (`fix(desktop): a project-tree root load only satisfies its own connection`) is new. Rebase drops the first two once #90334 merges.

With #90334 applied on current `main`, the right-rail file tree can still wedge on a **profile switch**. Reproduced on macOS 15 (Apple Silicon) at `a588685fb9`, desktop in global-remote mode (one `hermes serve` host, 5 profiles). Trace from the renderer (temporary `console.debug` in `loadRoot`, the reset, and the load effect), switching `engineer → default`:

```
4.20s RESET
4.22s loadRoot START req=5 force=false key=connection:<host>:engineer storeCwd=EMPTY   ← #90334 re-arm, under the OLD key
4.23s mainEffect connectionChanged=false last= key=connection:<host>:default            ← reset blanked lastConnectionKey
4.23s loadRoot SKIP loaded=false loading=true req=5                                     ← non-forced reload declined
4.94s loadRoot DROP req=5 latestReq=5 latestCwd=set keyNow=…:default keyThen=…:engineer ← stale-connection guard
```

`resetProjectTreeState()` runs on the gateway-scope change *before* the swapped `$connection` is published (`selectProfile` → `ensureGatewayProfile` resolves later), so the re-arm issues its read under the outgoing connection key. When the new descriptor lands, the load effect sees `lastConnectionKey === ''` (the reset blanked it), so it is not a `connectionChanged`, and its non-forced `loadRoot` returns early because req 5 is "in flight". Req 5 then completes and is — correctly — discarded by the `desktopFsCacheKey() !== connectionKey` guard from 9abeb89a49, which landed after #90334 was cut. Net: the store owns the cwd with `rootLoading` stuck `true`, Refresh is `disabled={loading}`, and the tree shows the skeleton until the next switch. It only bites when the read outlives the swap (occluded window → throttled timers, slow remote, large directory), which matches the "roughly 1 in 3 boots" wording in #90229.

**Fix:** record the connection key a root load was issued under in `ProjectTreeState`, and let `loadRoot`'s early return treat a load as "in flight" only when it is for the **same** key:

```ts
if (!force && current.cwd === cwd && current.connectionKey === connectionKey && (current.loaded || current.rootLoading)) {
  return
}
```

[...]

PR #97320, opened 2026-08-28
fix(desktop): reload project tree after profile reset

hailyuzairi:
## Summary
- reload the mounted project tree when an external profile/gateway reset invalidates its in-flight root read
- preserve request-generation checks so superseded directory responses cannot overwrite the replacement tree
- cover the profile-switch race and blank-CWD reset behavior with regression tests

## Root cause
During a live profile switch, the mounted tree can begin reading the new root before the parent profile-switch effect calls `resetProjectTreeState()`. The reset advances the root request ID, so the first response is correctly discarded, but the hook previously depended only on CWD and connection key. Neither changed again, leaving the right sidebar in its loading skeleton indefinitely.

## Verification
- `npm test -- --project ui src/app/right-sidebar/files/use-project-tree.test.ts` (20 passed)
- `npm test -- --project ui src/app/right-sidebar/index.test.tsx src/app/right-sidebar/files/use-project-tree.test.ts` (23 passed)
- `npm run test:ui` (6,135 passed)
- `npm run typecheck`
- `npm run lint` (0 errors; existing warnings only)
- `npm run build`
- independent code review: passed, no security or logic findings

### DIFF
--- a/apps/desktop/src/app/right-sidebar/files/use-project-tree.test.ts
+++ b/apps/desktop/src/app/right-sidebar/files/use-project-tree.test.ts
@@ -29,11 +29,16 @@ function ok(entries: { name: string; path: string; isDirectory: boolean }[]): He
 }
 
 describe('useProjectTree', () => {
-  it('starts empty when cwd is blank and skips IPC', async () => {
+  it('starts empty when cwd is blank and skips IPC across external resets', async () => {
     const { result } = renderHook(() => useProjectTree(''))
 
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: #90334 and #96329 address the same project-tree reset/skeleton latch with different re-arm mechanisms. This PR uses a reset generation dependency; please compare the approaches before choosing one.

Enough1122:
@D:\Hermes\tmp\review-body-97320.txt

PR #42262, opened 2026-06-08
fix(cron): inject chained response payloads

mijanx:
## Summary
- extract the useful `## Response` body from stored cron job output before injecting `context_from` prompt context
- fall back to `## Error` when a failed upstream job has no response section
- keep the existing 8K context clamp, but apply it after wrapper extraction so large prompt wrappers do not hide the actual result
- add regression tests for response extraction, error extraction, and truncation behavior

## Motivation
`context_from` lets one cron job use a previous job's output as context. Stored cron output markdown includes wrapper sections such as the original prompt, schedule metadata, and the final response. When the wrapper is large, downstream jobs can receive mostly prompt/history boilerplate while the useful response is truncated away.

This makes job chaining less reliable: downstream verifier/summarizer jobs should see the previous job's actual result, not the wrapper preamble.

## Test plan
- `git diff --check upstream/main..HEAD`
- `python -m py_compile cron/scheduler.py tests/cron/test_cron_context_from.py`
- `pytest tests/cron/test_cron_context_from.py -q -o 'addopts='` → 24 passed
- `env -u MATRIX_HOME_CHANNEL -u MATRIX_HOME_ROOM_ID -u MATRIX_HOME_CHANNEL_ID -u MATRIX_HOME_ROOM -u MATRIX_HOME_ROOM_THREAD_ID HERMES_HOME=$(mktemp -d) pytest tests/cron/test_cron_context_from.py tests/cron/test_scheduler.py tests/cron/test_cron_no_agent.py -q -o 'addopts='` → 176 passed

## Duplicate check
Searched open PRs/issues for:
- `context_from`
- `cron context_from`
- `context_from output`
- `recent completed output`

Nearest open PRs were cron test coverage and path hardening; neither changes `context_from` output extraction.

PR #97323, opened 2026-08-28
fix(cron): stop recursive boilerplate nesting in context_from continuity injections

ChrisTutorials:
## Problem

Cron jobs using `context_from` self-references (continuity mode) re-ingest their archived output each run. The archive is a wrapper:

```
# Cron Job: <name>
**Job ID:** ...
## Prompt
[the cron execution hint + full job prompt]
## Response
<the actual report>
```

The previous injection is *inside* that wrapper, so run N embeds run N-1's wrapper (including its "## Your previous run's output" block), which embeds run N-2's... The context block compounds recursively (we observed a 22 KB archive of which 19 KB was nested wrappers), the model sees stale header boilerplate every run, and it starts **mimicking the header format in its own reports** — delivering `# Cron Job:` / `Job ID:` / `## Prompt` slop to chat instead of a readable report.

## Fix

Extract the `## Response` section from the archive before injecting it as continuity context, and lower the cap from 8K to 4K chars (the extracted response is what matters; the wrapper was eating the budget). No behavioral change for jobs whose archives lack the wrapper (regex simply doesn't match).

## Testing

- `python3 -m pytest tests/cron -q`: **1033 passed, 1 skipped**.
- Verified against a live archive: 22,275-char wrapper → 1,947-char extracted response with the correct content.
- Field evidence: 6 recurring jobs on our install showed progressive header nesting in every continuity block before this change.

### DIFF
--- a/cron/scheduler.py
+++ b/cron/scheduler.py
@@ -4666,8 +4666,19 @@ def _build_job_prompt(
                 if not output_files:
                     continue  # silent skip — no output yet
                 latest_output = output_files[0].read_text(encoding="utf-8").strip()
-                # Truncate to 8K characters to avoid prompt bloat
-                _MAX_CONTEXT_CHARS = 8000
+                # The archive file is a wrapper ("# Cron Job … ## Prompt …
+                # ## Response …"). Injecting the whole wrapper back into the
+                # prompt compounds recursively: each run embeds the previous
+                # run's wrapper, which embeds the one before it, and models
+                # start mimicking the header format in their own reports.
+                # Extract the "## Response" section so continuity sees only
+                # what the job actually SAID last time.
+                _m = re.search(r"^## Response\s*\n\n?(.*)$", latest_output,
+                               re.S | re.M)
+                if _m:
+                    latest_output = _m.group(1).strip()
+                # Truncate to 4K characters to avoid prompt bloat
+                _MAX_CONTEXT_CHARS = 4000
                 if len(latest_output) > _MAX_CONTEXT_CHARS:
                     latest_output = latest_output[:_MAX_CONTEXT_CHARS] + "\n\n[... output truncated ...]"
                 if latest_output:


alt-glitch:

Duplicate of #42262. Both extract the archived response before context_from injection; #42262 is the broader canonical implementation with robust wrapper parsing, error fallback, and coverage.

gweber:
Reproduced and confirmed on another install (hermes-agent 0.20.4, aarch64,
Python 3.11). Three recurring jobs with `context_from: ["self"]`, measured by
calling `_build_job_prompt()` against the real archives — before and after:

| job | injected payload, before | first line, before | after | first line, after |
|---|---|---|---|---|
| openwein-tageslage | 8,028 | `# Cron Job: openwein-tageslage` | **3,490** | `**openwein.de — Tageslage, 30.08.2026**` |
| openwein-weingueter | 8,028 | `# Cron Job: openwein-weingueter` | **2,870** | `Runde abgeschlossen. Hier der Bericht …` |
| server-hygiene | 8,028 | `# Cron Job: server-hygiene` | **1,398** | `# Server-Hygiene 30.08` |

Before the patch all three inject an identical 8,028 characters of wrapper
header and prompt; after it, each injects its own report. `tests/cron`:
1033 passed, 1 skipped.

**The 4 K cap holds** on these archives — largest extraction is 3,490, i.e.
87 % of budget, and that is a job whose reports are deliberately long.

One amplifier that is not in the description yet: **a large attached skill
blows the wrapper before recursion contributes anything.**

| job | archive | `## Response` | signal | nesting depth |
|---|---|---|---|---|
| openwein-tageslage | 170,170 | 3,490 | 2.1 % | 1 |
| openwein-weingueter | 95,925 | 2,870 | 3.0 % | 1 |
| server-hygiene | 20,764 | 1,398 | 6.7 % | 2 |

`server-hygiene` matches your 22 KB / 19 KB case closely. The other two do
not: nesting depth is only **1**, yet the archive is 8× larger. The dominant
term there is the attached operator skill — 87,214 characters, present twice
in the archive. At that size the 8 K prefix is consumed by the skill's opening
paragraphs on the *first* continuity run. Depth grows the wrapper over time;
skill size makes it large immediately. Independent, and they compound.

One negative observation, in case it narrows the symptom: we did **not** see
the header-format mimicry. All three jobs kept producing clean reports and
never emitted `# Cron Job:` boilerplate downstream. If that turns out to be
model-dependent, mimicry would be a symptom of this bug rather than a reliable
detector of it — worth knowing for anyone trying to spot it in the wild.

What it cost us: the daily job's own handover — "stopped here, these leads are
still open" — never reached the next run, for six runs. Its operator
instructions even carried a line stating that the report *is* the handover.
It was not. We have disabled `context_from` on the affected jobs as an interim
measure and moved handover state into the per-job notepad, which is injected
uncapped.

PR #97327, opened 2026-08-28
fix(bedrock): recover from server-side cachePoint rejections per placement (Nova toolConfig.tools)

### LINKED ISSUE #97281: [Bug]: Nova models reject cachePoint in toolConfig.tools - ValidationException

JoaoMarcos44:
## What does this PR do?

Amazon Nova on the Bedrock Converse path fails **every tool-enabled turn** because Hermes puts a `cachePoint` block inside `toolConfig.tools`, which Nova rejects outright:

```
ValidationException: Malformed input request: #/toolConfig/tools/18:
extraneous key [cachePoint] is not permitted, please reformat your input and try again.
```

Bedrock's caching rules are **per-model-family AND per-field** — Nova accepts `cachePoint` in `system` and `messages`, but not in `toolConfig.tools`. `agent/bedrock_adapter.py` decided placement from a single static allowlist (`_CACHE_POINT_PATTERNS`) that only answers *"does this model cache at all"*, never *"in which section"*. Any family whose placement rules differ breaks 100% of turns until a maintainer edits the table and ships a release, with no runtime recovery and no user-side workaround — the exact maintenance trap the neighbouring `_NON_TOOL_CALLING_PATTERNS` comment already admits to (*"If a model fails with a tool-related ValidationException, add it here"*).

This PR makes **Bedrock's own verdict authoritative alongside the table**: classify the rejection from the JSON pointer AWS returns, drop the marker for that one section, retry the request once, and remember the verdict for the rest of the process so later turns are built clean. Every other section keeps its cache marker — Nova still gets `system` + `messages` caching instead of losing prompt caching wholesale.

This mirrors the module's existing self-heal idiom (`is_streaming_access_denied_error` → fall back to non-streaming `converse()`).

```mermaid
graph TD
    A[Converse Request] -->|Static Allowlist Verdict| B[Cache Markers Placed]
    B --> C{Bedrock Judgement}
    C -->|Accepted| D[Streamed Response]
    C -->|extraneous key cachePoint| E[Rejection Pointer toolConfig/tools]
    E --> F[Classify Placement]
    F --> G[Strip That Marker Only]
    G --> H[Retry Once - Same Turn]
    H --> D
    F --> I[Record Verdict Per Model]
    I -->|Next Turn Built Clean| B
    G -.->|System + Messages Caching Survives| D
```

## Related Issue

Fixes #97281

## Relationship to #97289

#97289 fixes the same crash from the **static-table side** — it splits an Anthropic-only tool-placement allowlist out of `_CACHE_POINT_PATTERNS`. This PR deliberately does **not** touch that allowlist; it adds the **runtime layer** underneath it, which nothing currently owns:

| | #97289 (static table) | This PR (server verdict) |
|---|---|---|
| Nova + tools today | fixed by hardcoded family list | fixed by rejection → strip → retry |
| A new family with different placement rules | breaks until a release edits the table | self-heals on the first request |
| Non-canonical model ids (proxies, custom inference profiles) | not matched by the pattern | classified from the server's own pointer |
| Sections other than the rejected one | unchanged | keep caching (no wholesale disable) |

The two stack cleanly and are complementary: with both, the static table avoids the round trip for known families, and the runtime layer catches everything the table has not learned yet. If #97289 lands first, the tool gate here becomes `_model_supports_tool_cache_point(model) and cache_point_allowed(model, "tools")` — a one-line merge.

## Type of Change

## Changes Made

- `agent/bedrock_adapter.py`
  - `cache_point_rejection_placement(exc)` — classifies a cachePoint ValidationException into `tools` / `system` / `messages` from the JSON pointer AWS returns; returns `None` for unrelated errors (including *other* `extraneous key` rejections such as `toolChoice`).
  - `strip_cache_points(kwargs, placement)` — returns a copy of the Converse kwargs with only that section's marker removed; returns the **input object unchanged** when there was nothing to strip, so callers can tell a retry is futile.
  - `note_cache_point_rejection` / `cache_point_allowed` / `reset_cache_point_rejections` — a per-process, per-model, per-placement verdict registry.
  - `recover_from_cache_point_rejection(exc, kwargs)` — the one call sites use: records the verdict, logs it once, and returns retry kwargs (or `None` → caller re-raises).
  - `build_converse_kwargs` now gates each of the three placements on `cache_point_allowed(...)`, so a learned verdict removes the marker at build time on every later turn.
  - Retry wired into `call_converse` and `call_converse_stream`.
- `agent/chat_completion_helpers.py`: same retry at both Bedrock dispatch sites — `_dispatch_nonstreaming_api_request` and `_open_bedrock_stream` (the streaming one is the path in the report; the retry stays inside the same managed Relay attempt, matching the IAM fallback next to it).
- `tests/agent/test_bedrock_adapter.py`: 11 tests in `TestCachePointRejectionRecovery` + an autouse fixture that keeps the process-wide registry from leaking between tests.

## Behaviour notes

- **Claude is untouched.** All three markers still go out for `anthropic.claude`, verified by an explicit test that a Nova rejection does not suppress Claude's tool marker.
- **No unbounded retry.** The retry fires only when a marker was actually removed; a cachePoint rejection with nothing to strip returns `None` and the original exception propagates.
- **Caching is not disabled wholesale** — that is the point of classifying by placement rather than flipping one global flag.
- No in-place mutation: `strip_cache_points` builds new dicts/lists and leaves the caller's kwargs intact.

## How to Test

1. `python -m pytest tests/agent/test_bedrock_adapter.py -q`
   - Observed: **76 passed** (65 pre-existing + 11 new).
2. Fail-closed verification of both layers (each reverted separately, then restored):
   - Removing the retry in `call_converse` → `test_call_converse_retries_without_the_marker` fails with the exact Nova `ValidationException` from the issue.
   - Removing the `cache_point_allowed(model, "tools")` gate in `build_converse_kwargs` → `test_recovery_records_verdict_so_later_turns_omit_the_marker` fails (the marker comes back on the next turn).
3. Bedrock surface: `python -m pytest tests/agent/test_bedrock_integration.py tests/agent/test_bedrock_empty_text_blocks.py tests/agent/test_bedrock_1m_context.py -q`
   - Observed: **40 passed**.
4. Transport/dispatch surface: `python -m pytest tests/agent/transports tests/agent/test_chat_completion_helpers_provider_sort.py -q`
   - Observed: **309 passed**.
5. `python -m ruff check agent/bedrock_adapter.py agent/chat_completion_helpers.py tests/agent/test_bedrock_adapter.py` → All checks passed.

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate — #97289 covers the static-table layer; this PR deliberately does not touch it (see the comparison above)
- [x] I've run the affected suites and all pass (numbers above)
- [x] I've tested on my platform: Windows 11 (x64), Python 3.14

- [x] I've updated relevant documentation (docstrings) — no user-facing docs change; the recovery is automatic
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — N/A, no new config keys
- [x] I've updated `CONTRIBUTING.md` / `AGENTS.md` if I changed architecture or workflows — N/A
- [x] I've considered cross-platform impact — N/A, pure wire-format logic

## For New Skills

N/A — not a skill.

## Screenshots / Logs

N/A — wire-format change covered by unit tests; the exact server error is quoted in #97281 and pinned verbatim in `NOVA_TOOLS_REJECTION`.

### DIFF
--- a/agent/bedrock_adapter.py
+++ b/agent/bedrock_adapter.py
@@ -645,6 +645,162 @@ def _model_supports_prompt_cache(model_id: str) -> bool:
     return any(pattern in model_lower for pattern in _CACHE_POINT_PATTERNS)
 
 
+# ---------------------------------------------------------------------------
+# Server-verdict cachePoint suppression
+# ---------------------------------------------------------------------------
+# The allowlist above is a static guess about *placement*, and Bedrock's real
+# rule is per-model-family AND per-field: Amazon Nova accepts cachePoint in
+# ``system``/``messages`` but rejects it inside ``toolConfig.tools`` with a
+# hard ValidationException that fails the whole request (#97281). Any static
+# table drifts the moment AWS ships a family whose placement rules differ, and
+# the failure mode is 100% of turns with no recovery and no user workaround.
+#
+# So the table is not the only authority: when Bedrock names a placement as
+# unpermitted, that verdict is recorded and the marker is dropped from that
+# placement for the rest of the process, and the rejected request is retried
+# once without it. Mirrors the existing self-heal idiom in this module
+# (is_streaming_access_denied_error → non-streaming converse()).
+
+CACHE_POINT_PLACEMENTS = ("tools", "system", "messages")
+
+# model_id (lowercased) → placements Bedrock has rejected this process.
+_CACHE_POINT_REJECTIONS: Dict[str, set] = {}
+
+# "#/toolConfig/tools/18: extraneous key [cachePoint] is not permitted"
+_CACHE_POINT_PATH_PATTERN = re.compile(
+    r"#/(?P<path>[A-Za-z0-9_./\[\]-]*)", re.IGNORECASE
+)
+
+
+def cache_point_rejection_placement(exc: BaseException) -> Optional[str]:
+    """Return the Converse section whose cachePoint block Bedrock refused.
+
+    Returns one of ``CACHE_POINT_PLACEMENTS``, or None when the error is not a
+    cachePoint rejection. Bedrock reports it as a ValidationException naming
+    the offending JSON pointer, e.g.::
+
+        Malformed input request: #/toolConfig/tools/18: extraneous key
+        [cachePoint] is not permitted, please reformat your input and try again.
+
+    Detection is message-based on purpose: the pointer is the only part of the
+    response that says *which* section was rejected, and the same wording
+    reaches us both as a raw botocore ``ClientError`` and wrapped by SDKs.
+    """
+    msg = str(exc)
+    lowered = msg.lower()
+    if "cachepoint" not in lowered:
+        return None
+    if "not permitted" not in lowered and "extraneous" not in lowered:
+        return None
+    match = _CACHE_POINT_PATH_PATTERN.search(msg)
+    path = (match.group("path") if match else "").lower()
+    if "toolconfig" in path or "tools" in path:
+        return "tools"
+    if "system" in path:
+        return "system"
+    if "messages" in path:
+        return "messages"
+    # A rejection we cannot localise: suppress the tool marker first, since
+    # toolConfig.tools is the only placement any supported family is known to
+    # refuse while still accepting the others.
+    return "tools"
+
+
+def note_cache_point_rejection(model_id: str, placement: str) -> None:
+    """Record that ``model_id`` refuses cachePoint blocks in ``placement``."""
+    if placement not in CACHE_POINT_PLACEMENTS:
+        return
+    _CACHE_POINT_REJECTIONS.setdefault(model_id.lower(), set()).add(placement)
+
+
+def cache_point_allowed(model_id: str, placement: str) -> bool:
+    """Return False once Bedrock has refused this placement for this model."""
+    return placement not in _CACHE_POINT_REJECTIONS.get(model_id.lower(), ())
+
+
+def reset_cache_point_rejections() -> None:
+    """Clear recorded cachePoint rejections. Used in tests."""
+    _CACHE_POINT_REJECTIONS.clear()
+
+
+def _is_cache_point_block(block: Any) -> bool:
+    return isinstance(block, dict) and set(block.keys()) == {"cachePoint"}
+
+
+def strip_cache_points(kwargs: Dict[str, Any], placement: str) -> Dict[str, Any]:
+    """Return a copy of Converse kwargs with ``placement``'s cachePoint removed.
+
+    Returns the input unchanged (same object) when there was nothing to strip,
+    which is what callers use to decide a retry cannot help.
+    """
+    if placement == "system":
+        system = kwargs.get("system")
+        if not isinstance(system, list):
+            return kwargs
+        cleaned = [b for b in system if not _is_cache_point_block(b)]
+        if len(cleaned) == len(system):
+            return kwargs
+        return {**kwargs, "system": cleaned}
+
+    if placement == "tools":
+        tool_config = kwargs.get("toolConfig")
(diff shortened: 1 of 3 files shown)


alt-glitch:

Related to #97289: both repair Nova's rejected tool-level cachePoint marker. This PR uses a runtime Bedrock-verdict retry and per-placement suppression; #97289 uses a static Anthropic-only tool-marker allowlist.

kshitijk4poor:
Merged via the P0 batch salvage PR #97704 (rebase merge — your commit landed on main with your authorship preserved).

https://github.com/NousResearch/hermes-agent/pull/97704

Thanks for the fix!

Issue #97107, opened 2026-08-28
[Bug]: MCP reconnect loop reuses startup-expanded ${ENV} headers — a rotated static token is never picked up automatically


adnankassem:
### Bug Description

For HTTP MCP servers configured with static header templates, e.g.

```yaml
mcp_servers:
  example:
    url: https://example.com/mcp
    headers:
      Authorization: "Bearer ${EXAMPLE_TOKEN}"
```

the `${ENV}` placeholder is resolved once, when the server task is built, and the resolved value is then reused by every automatic reconnect for the life of the process. If the underlying value is rotated (new token written to `~/.hermes/.env` or the process environment), the server task keeps presenting the stale header on every reconnect attempt and can never recover on its own.

Mechanism, verified at tag **v2026.8.27** (commit `5fc308a70719a83cccdbba4c0e39c23f5a8239d5`):

1. **Expansion happens at config load, not at connect.** `_load_mcp_config()` interpolates `${VAR}` / `${env:VAR}` into literal values (`tools/mcp_tool.py:5788`, interpolator at `tools/mcp_tool.py:5649`). A second expansion layer with the same timing exists in `load_config()` (`hermes_cli/config.py:2782`, applied at `hermes_cli/config.py:3893`; its cache is env-snapshot-invalidated per `hermes_cli/config.py:285-294`, but that only helps future config loads, not running tasks).
2. **The expanded dict is captured once per server task.** `MCPServerTask.run()` stores it (`tools/mcp_tool.py:3939`) and its reconnect loop (`tools/mcp_tool.py:4021`) passes the same start-time dict into every transport (re)build: `self._run_http(config)` at `tools/mcp_tool.py:4024`, `self._run_stdio(config)` at `tools/mcp_tool.py:4026`.
3. **Headers are rebuilt from that stale dict on every reconnect.** `_run_http()` does `headers = dict(config.get("headers") or {})` at `tools/mcp_tool.py:3599`. Nothing between reconnect cycles re-reads config or re-resolves placeholders.
4. **`reconnect_mcp_server()` takes no fresh input.** It only signals the existing task's `_reconnect_event` (`tools/mcp_tool.py:4770`), so the revived cycle reuses the same stale dict. Its only callers are OAuth-flow completion handlers (`hermes_cli/web_server.py:13876`, `tui_gateway/mcp_oauth_sessions.py:184`) — there is no trigger tied to env/config change.
[...]

Issue #78642, opened 2026-08-04
Shard tools/mcp_tool.py (god-file decomposition)


andrexibiza:
**Type:** Refactoring / Architecture
**Guideline Match:** "Refactor god-files into clean modules." / "Extend, don't duplicate. Design one shared interface."

**Description:**
`tools/mcp_tool.py` is a god file: **7,230 lines** on `main` (`42708f8bb3`). Repo-wide policy (2026-08): **all god files are sharded, never reverted** — the only correct answer is to shard the entire thing. Precedent: `gateway/run.py` (#54962) — sharding SHIPPED via 38 extraction PRs (all individually linked on the epic #78647). Nothing here is a request to revert, shrink-truncate, or abandon the file.

**Proposed Changes:**
- **Graph-gated decomposition (gate 1 first):** model the whole module (not pre-seeded clusters) with double-blind independent trees, merge to one typed graph under agreement provenance (≥4/5 agreed = load-bearing), then quality-gate the graph before any cut.
- **One PR per clean cluster/mixin, in waves:** byte-verbatim moves (docstrings, comments, exclusive constants), identity-preserving re-exports (`<module>.<name>` stays green), `Part of` linkage to this issue, suite-green on current main before open.
- **Zero behavior change.** No reverts. Full shard or untouched — partial "just extract this one class" escapes are not the policy.
- **Test contracts move with the code:** monkeypatch/logger patch targets and source-scanning tests that read this file get their targets updated in the same extraction PR.

**Tracking:** consolidated under the repo-wide god-file epic (meta issue, created with this series).

Part of #78647.

**Live collider interlock**
- #76586 — MCP session-liveness monitoring integration; conforming rebase-on-behalf run `rebase-76586` is IN FLIGHT at source `537f2247bdf4d56ee170ab339140e644ec11c898` onto target `c0106e50e7ecedb3ce34e785d949725dc4e0e457`. No delivery or adoption claim exists until 5×2×3 adjudication and two independent reviewers pass.

andrexibiza:
Tracking epic: #78647 — repo-wide god-file sharding (20 files, this file included). Scoreboard updates land here per shipped slice.

andrexibiza:
# KILL LOCK — tools/mcp_tool.py

**God file:** `tools/mcp_tool.py`
**Shard issue:** #78642 (OPEN — zero staged work at kill-start)
**Campaign EPIC:** #78647 (repo-wide god-file sharding, 20 tracked gods)
[...]

Issue #78628, opened 2026-08-04
Shard hermes_cli/web_server.py (god-file decomposition)


andrexibiza:
**Type:** Refactoring / Architecture
**Guideline Match:** "Refactor god-files into clean modules." / "Extend, don't duplicate. Design one shared interface."

**Description:**
`hermes_cli/web_server.py` is a god file: **17,700 lines** on `main` (`42708f8bb3`). Repo-wide policy (2026-08): **all god files are sharded, never reverted** — the only correct answer is to shard the entire thing. Precedent: `gateway/run.py` (#54962) — sharding SHIPPED via 38 extraction PRs (all individually linked on the epic #78647). Nothing here is a request to revert, shrink-truncate, or abandon the file.

**Proposed Changes:**
- **Graph-gated decomposition (gate 1 first):** model the whole module (not pre-seeded clusters) with double-blind independent trees, merge to one typed graph under agreement provenance (≥4/5 agreed = load-bearing), then quality-gate the graph before any cut.
- **One PR per clean cluster/mixin, in waves:** byte-verbatim moves (docstrings, comments, exclusive constants), identity-preserving re-exports (`<module>.<name>` stays green), `Part of` linkage to this issue, suite-green on current main before open.
- **Zero behavior change.** No reverts. Full shard or untouched — partial "just extract this one class" escapes are not the policy.
- **Test contracts move with the code:** monkeypatch/logger patch targets and source-scanning tests that read this file get their targets updated in the same extraction PR.

**Tracking:** consolidated under the repo-wide god-file epic (meta issue, created with this series).

andrexibiza:
Tracking epic: #78647 — repo-wide god-file sharding (20 files, this file included). Scoreboard updates land here per shipped slice.

andrexibiza:
## web_server.py scoreboard — wave 1 complete (7 slices)

All 7 slices double-blind reviewed, blind-flagged defects fixed by fix lanes, class-fix (CI seam-test collection) coordinated across the wave. All on true upstream base (aec331899e).

| Slice | PR | Module | Delta | Evidence |
|---|---|---|---|---|
| R1-C9 model-assignment | #79123 | web_server_model_assignment.py | +2 names | 40 tests, runner 36/0 |
| R2-B gateway-topology | #79124 | web_gateway_topology.py | 10 names, -186 | 31 tests, C1 fix GREEN |
[...]

PR #74809, opened 2026-07-30
fix(mcp): isolate per-server env_file resolution

seppegadeyne:
## What does this PR do?

Adds isolated per-server `env_file` support to native MCP configuration and uses one resolution path for runtime discovery and CLI probes.

This preserves the original contribution from #53647 as commit `9403c19a7` and completes the review requests on that PR: runtime parity, explicit precedence, same-key profile/server coverage, and no process-global secret mutation.

## Related Issue

Closes #53646.

Related work checked before opening this PR:

- #53647 is the exact partial fix. This PR supersedes it while preserving Ahmad's original commit and authorship.
- #49306 passes an env-file pointer to catalog MCPs such as n8n; it is complementary and does not resolve per-server placeholders in Hermes.
- #57369 adds trusted project-local MCP definitions; it is broader project-scope plumbing and does not implement this `env_file` contract.
- #51472 validates unresolved placeholders at save time; it is complementary to runtime/probe resolution.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [x] 🔒 Security fix
- [x] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Added `mcp_servers.<name>.env_file` for stdio and HTTP servers.
- Resolves relative paths from a valid `TERMINAL_CWD`, otherwise from the Hermes process working directory; docs recommend absolute paths for gateways and cron jobs.
- Defines precedence for placeholders as per-server `env_file` → active profile secret scope / process environment.
- Keeps env-file values in an isolated mapping instead of writing them to `os.environ`, preventing sibling-server and cross-profile inheritance.
- Uses the same resolver for runtime discovery and `hermes mcp` probe paths.
- Warns and safely falls back when the file is missing or unreadable.
- Revalidates the effective server config after interpolation so an env file cannot hide blocked stdio command arguments.
- Redacts exact env-file values from runtime discovery logs/status plus CLI, dashboard, OAuth, and catalog probe failures; configured URLs are not echoed in non-MCP endpoint errors.
[...]

PR #97331, opened 2026-08-28
feat(mcp): service-account M2M OAuth provider for HTTP MCP servers

jameswynn:
## Summary

- Adds a generic `auth: service_account` mode for HTTP MCP servers that authenticate machine identities (M2M) without browser interaction
- A long-lived service-account password (read from an environment variable at runtime, never from config) is exchanged for a short-lived Bearer token via the OAuth 2.0 `client_credentials` grant, injected into MCP requests, and renewed automatically
- This is **distinct** from `auth: oauth` (browser PKCE) and `headers:` (static key) — no user interaction required, and tokens are not statically embedded

## Config shape

```yaml
mcp_servers:
  toolhive:
    url: https://mcp.example.com/mcp
    auth: service_account
    service_account:
      token_url: https://idp.example.com/application/o/toolhive/token/
      client_id: toolhive
      username: zug
      password_env: AUTHENTIK_ZUG_APP_PASSWORD   # env-var NAME only — value in .env
      scope: "openid profile groups toolhive-audience"
      client_secret_env: OPTIONAL_CLIENT_SECRET   # optional
```

## Files changed

- **`tools/mcp_service_account.py`** (new): `ServiceAccountAuth` — an `httpx.Auth`-compatible provider implementing token exchange, disk cache (mode 0600, atomic write), proactive renewal (60 s buffer), one 401 retry, refresh-token path with SA-exchange fallback, per-server `asyncio.Lock` concurrency deduplication, and strict no-secrets-in-logs/errors policy
- **`tests/tools/test_mcp_service_account.py`** (new): 30 focused tests — first acquisition & header injection, cache reuse, near-expiry proactive renewal, refresh-token path and fallback, 401 retry, concurrent deduplication (one exchange for 5 concurrent requests), missing env var error, malformed response error, token endpoint 500 error, no password/token in logs, profile isolation, 0600 cache permissions, header/oauth backwards compat
- **`tools/mcp_tool.py`**: `elif self._auth_type == "service_account":` branch alongside existing `oauth` branch
- **`hermes_cli/mcp_config.py`**: `service_account` branch in `cmd_mcp_add` (validates config, surfaces missing env-var names without storing secrets); SA token cache cleanup in `cmd_mcp_remove`
- **`website/docs/reference/mcp-config-reference.md`**: full docs — Authentik example, sub-key reference table, behavior notes, comparison table vs `oauth` and `headers` modes

## Test results

```
191 passed, 2 deselected in 3.00s
```

The 2 deselected are pre-existing `TestCallbackPortReservation` failures in `test_mcp_oauth.py` (unrelated to this PR — a `result.code` attribute issue on the legacy tuple return from `_wait_for_callback`).

## Design tradeoffs

- **Generic, not Authentik-specific**: works with any IdP that accepts `client_credentials` + `username`/`password` on the token endpoint (Authentik, Keycloak, Auth0 machine-to-machine, etc.)
- **`password_env` only for v1**: extensible to `password_command` (exec without shell) without breaking the config schema — just add a new key; the resolver is a single function
- **No shell eval by default**: a `password_command` extension would exec without a shell and document the security implications; no generic arbitrary shell evaluation is added here
- **Token cache separate from OAuth cache**: uses `<server>-sa.json` suffix so SA and browser-OAuth tokens for the same server name don't collide
- **In-process dedup only**: the `asyncio.Lock` prevents concurrent exchanges within a process; cross-process dedup (e.g. file locking) is left for a future follow-up as it adds complexity for a rare edge case

### DIFF
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -1239,7 +1239,7 @@ def _is_env_config_key(key: str) -> bool:
     ]
     return (
         key_upper in api_keys
-        or key_upper.endswith(('_API_KEY', '_TOKEN', '_SECRET'))
+        or key_upper.endswith(('_API_KEY', '_TOKEN', '_SECRET', '_PASSWORD'))
         or key_upper.startswith('TERMINAL_SSH')
     )
 
--- a/hermes_cli/mcp_config.py
+++ b/hermes_cli/mcp_config.py
@@ -534,6 +534,31 @@ def cmd_mcp_add(args):
                 _info("Cancelled.")
                 return
 
+    elif url and auth_type == "service_account":
+        # Service-account M2M: validate the sub-block; secrets stay in env vars.
+        print()
+        _info(f"Configuring service-account auth for '{name}'...")
+        sa_cfg = server_config.get("service_account") or {}
+        from tools.mcp_service_account import validate_service_account_config
+        sa_errors = validate_service_account_config(name, sa_cfg)
+        if sa_errors:
+            for err in sa_errors:
+                _warning(err)
+            _warning(
+                "Fix the service_account block in config.yaml and re-run. "
+                "Secrets must be in environment variables, not in config."
+            )
+            return
+        server_config["auth"] = "service_account"
+        _success(
+            f"Service-account auth configured — token will be acquired from "
+            f"{sa_cfg.get('token_url', '(token_url)')} on first connection"
+        )
+        _info(
+            f"Ensure ${sa_cfg.get('password_env', 'PASSWORD_ENV')} is set "
+            "in your shell or in $HERMES_HOME/.env before connecting."
+        )
+
     elif url:
         # Prompt for API key / Bearer token for HTTP servers
         print()
@@ -668,6 +693,13 @@ def cmd_mcp_remove(args):
         from tools.mcp_oauth_manager import get_manager
         get_manager().remove(name)
         _success("Cleaned up OAuth tokens")
+    except Exception:
+        pass
+
+    # Clean up service-account token cache if present.
+    try:
+        from tools.mcp_service_account import remove_service_account_tokens
+        remove_service_account_tokens(name)
     except Exception:
         pass
 
--- a/hermes_cli/mcp_config.py
+++ b/hermes_cli/mcp_config.py
@@ -803,6 +803,21 @@ def cmd_mcp_test(args):
     headers = cfg.get("headers", {})
     if auth_type == "oauth":
         _info("Auth: OAuth 2.1 PKCE")
+    elif auth_type == "service_account":
+        sa = cfg.get("service_account") or {}
+        token_url = sa.get("token_url", "(token_url not set)")
+        client_id = sa.get("client_id", "?")
+        username = sa.get("username", "?")
+        password_env = sa.get("password_env", "?")
+        _info(f"Auth: service account (M2M OAuth)")
+        _info(f"  token_url:    {token_url}")
+        _info(f"  client_id:    {client_id}")
+        _info(f"  username:     {username}")
+        _info(f"  password_env: {password_env}  (value read from env at connect time)")
+        if sa.get("scope"):
+            _info(f"  scope:        {sa['scope']}")
+        if sa.get("client_secret_env"):
+            _info(f"  client_secret_env: {sa['client_secret_env']}  (value read from env)")
     elif headers:
         for k, v in headers.items():
             if isinstance(v, str) and ("key" in k.lower() or "auth" in k.lower()):
--- a/hermes_cli/web_models.py
+++ b/hermes_cli/web_models.py
@@ -424,10 +424,13 @@ class MCPServerCreate(BaseModel):
     args: List[str] = []
     # env: KEY=VALUE map for stdio servers (API keys, etc.)
     env: Dict[str, str] = {}
-    # auth: "none" | "oauth" | "header" | None
+    # auth: "none" | "oauth" | "header" | "service_account" | None
     auth: Optional[str] = None
     # One-time provisioning input; persisted only to the profile's .env.
     bearer_token: Optional[SecretStr] = None
+    # Non-secret service-account fields (token_url, client_id, username, scope,
+    # password_env, client_secret_env). Secrets are referenced by env-var name only.
+    service_account: Optional[Dict[str, Any]] = None
     profile: Optional[str] = None
 
 
--- a/hermes_cli/web_server.py
+++ b/hermes_cli/web_server.py
@@ -13678,7 +13678,7 @@ def _normalize_mcp_server_create(
 
     if bool(url) == bool(command):
         raise ValueError("Provide exactly one of URL (HTTP/SSE) or command (stdio)")
-    if auth not in {"none", "header", "oauth"}:
+    if auth not in {"none", "header", "oauth", "service_account"}:
         raise ValueError(f"Unsupported auth mode: {auth}")
 
     server_config: Dict[str, Any] = {}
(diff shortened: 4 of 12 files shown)


andrexibiza (commented):
Reviewed exact head `85625bc349ba97863cda1d70355964e9a7278609`. I also reconciled it against live `main@6da0ae1cf5a37898a046b644cf23f9fe67baba22`: the branch is currently 3 commits ahead / 226 behind from merge base `f3cbb262c1d7014f1a4d225242d06983122c5ddf`, so the final shape needs a current-main restack before it can be treated as a landing object.

There is a lot worth keeping here: the auth provider is already a bounded module rather than another giant inline implementation; disk-cache writes are atomic and `0600`; error paths avoid echoing token response bodies; the SDK-httpx compatibility handling is careful; the one-401 retry is bounded; and the test surface covers cache reuse, renewal, refresh fallback, concurrency, profile isolation, and legacy auth coexistence. I found no prior review/comment on this head, so this is not repeating an unchanged-head review.

I do see two P1 correctness/security blockers plus a landing-order gate.

### P1 — the configured token endpoint is not the actual credential boundary

`validate_service_account_config()` currently accepts both `https://` **and `http://`** token URLs:

```py
if token_url and not str(token_url).startswith(("https://", "http://")):
```

That directly contradicts this PR's own config reference, which says `token_url` is `https://`-required, and it matters because this request contains a long-lived service-account password and optionally a client secret. OAuth 2.0 requires TLS for password-based client authentication; this path cannot make plaintext credential transport a supported configuration.

There is a second side of the same boundary: the dedicated token client is created as `AsyncClient(follow_redirects=True)`. A 307/308 is method-preserving, so an authorization server (or compromised/misconfigured endpoint) can redirect the POST and its credential-bearing form to a different `Location`. The config proves one intended secret sink (`token_url`), but runtime silently widens that proof to whatever redirect target is returned. That is exactly the kind of credential egress boundary that should fail closed.

Required shape: require HTTPS at validation and consumption, and do **not** automatically follow token-endpoint redirects. If redirect support is genuinely required, it needs an explicit same-origin + HTTPS policy before replaying a credential-bearing body; cross-origin redirects must not receive the password/client secret. Please add adversarial regressions for (1) `http://idp/...` refusal and (2) a 307/308 to a different origin proving the form is not replayed. RFC 6749 §2.3.1/§4.4.2 and RFC 9110 §15.4.8/§15.4.9 are the relevant protocol contracts.

### P1 — the public contract says generic OAuth client credentials, but the wire contract is an Authentik extension

The PR body and docs explicitly say this is “generic, not Authentik-specific” and name Authentik, Keycloak, and Auth0 M2M. The implementation, however, **requires** `username` + `password_env` and emits:

```py
{
    "grant_type": "client_credentials",
    "client_id": ...,
    "username": ...,
    "password": ...,
}
```

That exact request is valid for Authentik's documented service-account extension. It is not the generic OAuth 2.0 client-credentials request. RFC 6749 §4.4.2 defines the request body as `grant_type=client_credentials` plus optional `scope`, with the confidential client authenticating to the token endpoint. Auth0 M2M uses client credentials and generally an `audience`; Keycloak's service-account client-credentials flow is likewise client authentication, not a required resource-owner username/password pair. The current schema has no `audience` and makes `username`/`password_env` mandatory, so the advertised provider parity is not supported by this implementation.

This can be repaired cleanly in either direction:

1. Narrow the feature contract to the Authentik-style app-password exchange it actually implements, including naming/docs/tests; or
2. Make the grant strategy typed: a standards-compliant `client_credentials` mode with actual client authentication (and extensible provider parameters such as `audience`), plus an explicit Authentik service-account/app-password variant. Do not infer provider semantics from the presence of free-form fields.

Please add at least one standards-conforming client-credentials witness and one Authentik app-password witness. Right now the tests prove the implementation reproduces its own form, not that the advertised protocol surface interoperates.

### Credential lifecycle interlock

This also composes with #97107 and #74809 rather than existing in isolation. #97107 establishes that automatic MCP reconnect currently reuses startup-resolved credential state; this provider reads `os.environ` on each exchange, which is better, but the docs say `$HERMES_HOME/.env` is loaded before MCP connection. Editing/rotating the profile `.env` does not itself mutate an already-running process environment, so an unattended password rotation can still leave renewal/reconnect presenting stale authority unless the scoped secret source is reloaded. #74809 (`seppegadeyne`, preserving the earlier #53647 contribution) is complementary per-server secret-scope work and specifically avoids process-global secret mutation. Please settle one secret-resolution owner/precedence when rebasing rather than creating a second credential sourcing model. If live `.env` rotation is intentionally out of scope, state the restart/reload requirement explicitly; do not imply automatic renewal also means automatic source-secret rotation.

### Landing / ownership gate

The branch is 226 commits behind current main and adds behavior directly into two live god-file fracture targets:

- `tools/mcp_tool.py` — #78642 (7,230-line decomposition owner)
- `hermes_cli/web_server.py` — #78628 (17,700-line decomposition owner)

The good news is that `tools/mcp_service_account.py` is already the obvious bounded owner. Keep that. On the restack, the integration should remain a thin seam and the dashboard/config behavior should compose with the current extracted MCP/web routing owners rather than regrowing the old files. There is active same-surface work across the MCP lifecycle and credential boundary (#97107, #74809, plus the #78642 shard set), so merge order needs to be explicit after rebase rather than trusting a 226-commit-old hunk context.

### Acceptance / attribution

All three commits currently lack hosted-green acceptance: exact-head CI `33204357683`, Docker `33204356711`, and Nix `33204356799` are `action_required`, with CI reporting zero jobs; the preceding two commit heads show the same `action_required` pattern. Local `191 passed` is useful development evidence, but it is not an exact-commit hosted acceptance receipt.

The three commits are authored as `James Wynn <james@jameswynn.com>`. Current main has no contributor mapping for either that email or `jameswynn`, so attribution also needs to be made release-resolvable before landing. The `Claude Sonnet 4.6 <noreply@anthropic.com>` co-author trailer is already covered by the repository's explicit AI-assistant ignore policy, so that is not an attribution blocker.

This is a useful capability and the implementation already has several strong local properties. The important correction is to make the public protocol and the secret-egress boundary exactly as strong as the code claims: one exact credential destination, one explicit grant strategy, one scoped secret source, and green evidence on every landed commit. Once those are true, the core provider is in good shape to carry forward. 🚀

jameswynn:
Superseded by #97628, which contains every commit from this branch plus the fixes for both P1 blockers raised in review.

Addressed there:

**P1 — token endpoint was not the actual credential boundary.** `token_url` now requires `https://`, enforced at validation *and* re-checked in `_post_token_request` immediately before the credential-bearing body leaves the process, so a config that bypassed validation cannot slip through. The token client is built with `follow_redirects=False`, and any 3xx from the token endpoint is surfaced as an error rather than replaying the form at the redirect target. Regressions cover `http://` refusal, 301/302/303/307/308 non-replay (asserting exactly one request was issued), the error leaking neither password nor Location, and the live auth flow's client construction. Both assertions were mutation-checked — reverting either fix fails a test.

**P1 — generic OAuth claim vs. Authentik wire contract.** You were right that the emitted form is Authentik's service-account extension, not RFC 6749 §4.4.2. Narrowed to what is actually implemented: `service_account.grant_type` is now required and enum-checked, with `authentik_app_password` the only accepted value, and required fields resolved per strategy rather than inferred from field presence. The config-level strategy name is kept distinct from the wire `grant_type` it maps to, since Authentik reuses the `client_credentials` wire name. Docs now state plainly that Keycloak service accounts and Auth0 M2M are not supported here; adding a standards-conforming mode later is additive.

Also documented the credential-rotation interlock you raised: the password is re-read per exchange, but editing `$HERMES_HOME/.env` does not mutate a running process's environment, so rotation requires a restart — renewal renews the access token, not the source credential.

On the landing-order note: #97628 is now rebased onto current main and is 0 commits behind. Worth flagging that the decomposition owners you referenced have not landed yet — on current main `hermes_cli/web_server.py` is still ~813 KB and `tools/mcp_tool.py` ~386 KB, so there are no extracted MCP/web routing owners to compose with. The integration remains a thin seam (10 lines in `mcp_tool.py`, 32 in `web_server.py`) with `tools/mcp_service_account.py` as the bounded owner.

Closing this in favour of #97628 so there is one review surface and one landing object.

Issue #80213, opened 2026-08-06
[Bug] Desktop: Default project directory (project-dir.json) ignored on Windows — cwd always falls back to home


ZAKK-yrt:
## Bug Description

Hermes Desktop v0.20.0 (2026.8.3) on Windows 10 的"默认项目目录"(Default project directory) 设置不生效。

- `project-dir.json` 已正确设置为 `{"dir": "D:\\hermes coding"}`
- `terminal.cwd` 也配置为 `D:\hermes coding`
- 环境变量 `TERMINAL_CWD=D:\hermes coding` 已正确设置
- App 已完全退出并重启多次
- 但新会话 `pwd` 始终返回 `C:\Users\<user>`（home directory）

## Steps to Reproduce

1. 在 Hermes Desktop Settings → 会话 → 默认项目目录，设为 `D:\hermes coding`
2. 完全退出 app 并重启
3. 创建新会话 (Ctrl+N)
4. 执行 `pwd`
5. 结果：`/c/Users/<user>`（而非配置的目录）

## Expected Behavior

新会话的 working directory 应为 `D:\hermes coding`

## Actual Behavior

新会话 cwd 始终为 home directory (`C:\Users\<user>`)

## Diagnostic Evidence

DavidMetcalfe:
## Findings: not fixed yet, but a fix is in flight — this issue is a duplicate of #71873

Thanks for the detailed repro — the diagnostic evidence is exactly what we needed to pin this down. I investigated both this issue and #71873 against current `main`, and they're the same bug.

**Mechanism (verified in source):**

1. **New-chat cwd inherits the focused session's workspace.** `resolveNewSessionCwd()` (store/projects.ts) gained a "focused session workspace" step in commit `883076ccd7` (2026-08-01, "⌘N/⌘T keep the focused session's project"). After a restart the focused session is the just-resumed one, whose stored cwd is a stale fallback (home dir) — so Ctrl+N inherits home and never consults the configured default project dir. The `project-dir.json` / `terminal.cwd` / `TERMINAL_CWD` settings are all applied correctly; the renderer just never asks for them on this path.

[...]

Issue #40334, opened 2026-06-06
Desktop: profile switch ignores terminal.cwd, uses wrong profile's working directory


bigleon2026:
## Bug Description

When switching between profiles in Hermes Desktop, the `terminal.cwd` configured in the target profile is ignored. Instead, the terminal session inherits the cwd from a previously-used profile.

## Environment

- **Hermes version**: 0.16.0
- **OS**: Windows 10
- **Desktop App**: Hermes Desktop (native Electron app)

## Steps to Reproduce

1. Have two profiles with different `terminal.cwd`:
   - Profile A (`mahjong`): `terminal.cwd: D:\Cursor project\Mahjong`
   - Profile B (`ef-design`): `terminal.cwd: D:\P4-ProjectEF\Design\策划文档md`
2. Use Profile A in Desktop first (works correctly).
3. Switch to Profile B in Desktop.
4. Start a new session (`/reset`).
5. Ask the agent: "What is the current working directory?"
6. Run `pwd` in terminal.

## Expected Behavior

Profile B's terminal session uses `D:\P4-ProjectEF\Design\策划文档md` as configured.

## Actual Behavior

The terminal session uses Profile A's cwd (`D:\Cursor project\mahjong`). Earlier sessions (before `persistent_shell` was configured) fell back to the default profile's cwd (`C:\Users\shenlihang`).

## Evidence from agent.log

Profile B's agent.log shows correct `.env` loading but wrong cwd:

```
13:44:03 Loaded environment variables from ...\profiles\ef-design\.env
13:47:09 Session snapshot created (session=e984f874882b, cwd=D:\Cursor project\mahjong)
```

The cwd `D:\Cursor project\mahjong` belongs to Profile A.

For comparison, Profile A's agent.log correctly uses its own cwd:
```
13:24:48 Session snapshot created (session=ab83d9e087ea, cwd=D:\Cursor project\mahjong)
```

## Verified Config

Profile B's `config.yaml` terminal section is identical to Profile A's (except cwd), and `hermes config show --profile ef-design` reports the correct `Working dir: D:\P4-ProjectEF\Design\策划文档md`. The active_profile file also correctly points to `ef-design`.

## Workaround

Setting `terminal.shell_init_files` to a script that does `cd` on init works as a temporary fix, but the root cause should be addressed.

Omnichannel-nz:
[...]

PR #97334, opened 2026-08-28
fix(desktop): seed new-chat workspace from the focused profile's terminal.cwd

TomSpoct:
## Problem

In app-global remote mode, one backend serves every profile. Each profile can declare its own workspace via `terminal.cwd` in its `config.yaml`, and the gateway already honors it for session fallbacks (`_profile_configured_cwd`, added for the #40334 'rubberband to launch profile' bug).

The desktop, however, short-circuits that:

1. `seedDefaultCwd()` fetched `/api/fs/default-cwd`, which resolves the **launch** profile's `terminal.cwd` (or `TERMINAL_CWD` / process cwd), and wrote the result with `setCurrentCwd` — which **persists to localStorage** as if the user had chosen it. A single boot from the wrong folder poisoned the remembered workspace for that connection+profile permanently.
2. `ensureDefaultWorkspaceCwd()` in remote mode seeded only from the remembered value and never consulted the focused profile's `terminal.cwd`.

Net effect: a profile with `terminal.cwd: ~/workspaces/lex` opened every new chat in a stale folder (e.g. another agent's workspace), with no way to fix it permanently except manual re-picking.

## Fix

**Backend** (`hermes_cli/web_server.py`):
- `GET /api/fs/default-cwd` accepts an optional `?profile=` param and resolves that profile's `terminal.cwd` from its **own** `config.yaml` — read via `read_user_config_raw` + managed overlay + `${VAR}` expansion, mirroring the gateway's `_profile_configured_cwd` read-side pipeline. Unknown profiles return 404 (existing `_resolve_profile_dir`); profiles with no usable `terminal.cwd` fall back to the launch default, so behavior for unconfigured profiles is unchanged.
- No param → exactly the previous behavior.

**Desktop**:
- `desktopDefaultCwd()` now requests the **active profile's** default (`?profile=<desktopFsProfile()>`).
- `ensureDefaultWorkspaceCwd()` accepts an optional `remoteDefault` (fetched by the caller to avoid an import cycle) and prefers it over the remembered path in remote mode.
- `seedDefaultCwd()` now seeds **transiently** (`setCurrentCwdTransient`) instead of `setCurrentCwd`: derived defaults are no longer persisted as user choices, breaking the stickiness loop (#80213-class).

## Result

With `terminal.cwd` set in a profile's config, every new session of that profile opens in its own workspace with zero manual action — the behavior requested for per-agent workspaces. Launch-profile and no-config setups behave exactly as before.

## Testing

- `apps/desktop`: 17/17 tests pass in `desktop-fs.test.ts` (14 pre-existing updated for the new query param + 3 new: profile-scoped request, no-profile legacy path, special-char encoding). `tsc -p tsconfig.json --noEmit` clean.
- Backend: live route checks — `profile=lex` returns the profile's configured workspace, no param preserves legacy behavior, unknown profile → 404.
- `use-session-actions.test.tsx` failures (83) are pre-existing on upstream main (verified via `git stash` — identical failures without this change).

### DIFF
--- a/apps/desktop/src/app/gateway/hooks/use-gateway-boot.ts
+++ b/apps/desktop/src/app/gateway/hooks/use-gateway-boot.ts
@@ -62,6 +62,7 @@ import {
   setConnection,
   setCurrentBranch,
   setCurrentCwd,
+  setCurrentCwdTransient,
   setSessionsLoading
 } from '@/store/session'
 import {
@@ -550,16 +551,19 @@ export function useGatewayBoot({
     // Seed the working dir from the backend default on a fresh view (nothing
     // open yet). Shared by boot + soft switch.
     async function seedDefaultCwd(shouldPublish: () => boolean = () => true) {
-      await ensureDefaultWorkspaceCwd(shouldPublish)
+      const remoteDefault = await desktopDefaultCwd().catch(() => null)
+      await ensureDefaultWorkspaceCwd(shouldPublish, remoteDefault?.cwd ?? null)
 
       if (!shouldPublish()) {
         return
       }
 
-      const remoteDefault = await desktopDefaultCwd().catch(() => null)
-
       if (shouldPublish() && remoteDefault?.cwd && !$activeSessionId.get() && !$currentCwd.get()) {
-        setCurrentCwd(remoteDefault.cwd)
+        // Transient seed: the backend default is derived per-profile config,
+        // not a user choice. Persisting it (setCurrentCwd) promoted whatever
+        // folder the app launched in into the remembered workspace, and every
+        // later new chat inherited it (#80213-class stickiness).
+        setCurrentCwdTransient(remoteDefault.cwd)
         setCurrentBranch(remoteDefault.branch || '')
       }
     }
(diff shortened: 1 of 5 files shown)


Enough1122:

**Overall:** Fixes #80213-class workspace stickiness where the backend-derived default cwd was persisted as remembered workspace and inherited by every later new chat.

**What it does**
- `apps/desktop/src/app/gateway/hooks/use-gateway-boot.ts:seedDefaultCwd` reorders `desktopDefaultCwd()` before `ensureDefaultWorkspaceCwd` and threads `remoteDefault?.cwd` into it; the final publish uses `setCurrentCwdTransient(remoteDefault.cwd)` instead of `setCurrentCwd` — transient seed is not persisted as user choice. Second call site (post-connect) does the same `profileDefault` plumbing.
- Transient vs persisted distinction preserves user-chosen cwd while still seeding new-chat workspace from focused profile's `terminal.cwd` per profile config.

**Non-blocking notes**
- `desktopDefaultCwd().catch(() => null)` silently swallows errors — add debug log if profile config default consistently fails, to avoid silent empty workspace.
- Verify `ensureDefaultWorkspaceCwd` correctly treats `null` hint as no-op (it does — keeps remembered cwd as fallback).

No blocking issues.

PR #96272, opened 2026-08-27
fix(desktop): guard project mutations in all-profiles view

marcob896:
## Problem

The aggregated **All profiles** project tree exposes project actions whose RPCs require a specific profile. `projectParams()` rejects those writes before they reach the gateway, so Delete appears to do nothing after the optimistic rollback; Rename, Appearance, Add folder, and Set active have the same ownership problem.

## Fix

- derive a shared `canManage` guard from `$profileScope !== ALL_PROFILES`
- hide profile-owned identity and Appearance actions in the aggregated view
- replace Delete with a disabled prompt to select a profile
- preserve read-only path actions and local auto-project dismissal
- apply the guard consistently to both the kebab menu and context menu

## Testing

- `npx vitest run --project ui src/app/chat/sidebar/projects` → 5 files, 82 tests passed
- `npx eslint src/app/chat/sidebar/projects/project-menu.tsx src/app/chat/sidebar/projects/project-menu.test.tsx` → clean
- `npx tsc -p . --noEmit` → clean
- `git diff --check origin/main...HEAD` → clean

## Scope

This is the class-level project-action guard related to #94738. PR #94739 covers the Delete symptom and also adds a separate project-creation affordance; this PR deliberately keeps creation UX out and guards every profile-owned mutation exposed by both project menus.

## AI assistance disclosure

Root-cause analysis, implementation, and verification were performed with Hermes Agent. All reported commands were run against this branch.

PR #94739, opened 2026-08-25
fix(desktop): always offer project creation; disable project delete in all-profiles view

danieljustus:
Fixes #94738

## Problem

Two user-visible gaps in the projects sidebar, reported as "delete does nothing" and "no way to create a project":

1. **Delete can fail silently.** `projectParams()` in `store/projects.ts` throws whenever `$profileScope === ALL_PROFILES` ("Projects are unavailable while viewing all profiles"). `deleteProject()` removes the project optimistically, `persistOrRollback` restores it and re-throws — a confirmed delete looks like a no-op. Verified against the live backend that `projects.delete` itself works; the failure is purely the renderer scoping guard.
2. **Create is unreachable in the flat sessions view.** The section-header `+` in `app/chat/sidebar/index.tsx` only calls `openProjectCreate()` while `agentsGrouped`; otherwise the same button starts a session (`onNewSessionInWorkspace`). The shipped-but-invisible "New project" dialog (#54317) stays hidden for every user in the flat/ungrouped view.

## Fix

- `app/chat/sidebar/index.tsx`: render an additional `+` ("New project", `openProjectCreate()`) in the section header whenever `!showAllProfiles && !agentsGrouped` — exactly the mode in which the existing `+` is a new-session button, so no duplicate affordance in grouped mode. Reuses the existing `s.projects.newButton` i18n key (`New project`, already localized in all six locales).
- `app/chat/sidebar/projects/project-menu.tsx`: disable the destructive **Delete…** item for explicit projects while `$showAllProfiles` is active (project RPCs are per-profile and cannot be scoped with no active profile), instead of a confirmed delete that silently rolls back. Auto-project "Remove from sidebar" is unaffected (local dismissal).

## Testing

- `project-menu.test.tsx`: two new cases — Delete stays enabled outside the all-profiles view, and is `aria-disabled` while viewing all profiles (steerable `$showAllProfiles` mock).
- `npx vitest run src/app/chat/sidebar/projects/` → 5 files, **81 passed**.
- `npx tsc -p tsconfig.json --noEmit` → clean.

## Notes

- No behavior change for grouped mode (`agentsGrouped`): the original `+` already opened the create dialog there; the new button only covers the flat view.
- The all-profiles view already hides the create `+` (project RPCs are unavailable there); the disabled Delete is the consistent counterpart.

## AI assistance disclosure

Investigation (live WebSocket RPC repro against the local gateway, code audit) and the fix/test draft were done with the help of an autonomous coding agent. All commands shown were actually run.

PR #97335, opened 2026-08-28
fix(desktop): repair all-profile project actions

Ed3scomb3s:
## Summary

- preserve the concrete project id and profile behind each merged all-profiles project row
- route Delete to every explicit source record and reconcile the merged tree safely
- hide owner-specific writes and the blank-state New project action when browsing All profiles

## Root cause

The cross-profile tree intentionally merges projects by folder, but discarded the source project identity. The renderer then tried to mutate the merged row through the All profiles sentinel, which the project RPC guard correctly rejects because it is not a writable profile.

## Validation

- 46 focused desktop UI tests
- 8 cross-profile backend tests
- desktop TypeScript typecheck
- targeted ESLint and Ruff checks
- manual packaged-app verification: deleting a saved project removed its Hermes record while leaving its folder and repository untouched

### DIFF
--- a/apps/desktop/src/app/chat/sidebar/index.tsx
+++ b/apps/desktop/src/app/chat/sidebar/index.tsx
@@ -1900,7 +1900,9 @@ export function ChatSidebar({
           </div>
         )}
 
-        {!showSessionSections && <SidebarBlankState onNewProject={openProjectCreate} />}
+        {!showSessionSections && (
+          <SidebarBlankState onNewProject={showAllProfiles ? undefined : openProjectCreate} />
+        )}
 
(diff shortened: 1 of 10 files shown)


alt-glitch:

Related: #96272 and #94739 address the same all-profiles project-action cluster by hiding or disabling unroutable mutations. This PR instead preserves source identities and routes Delete to each owning record; maintainers should choose the desired all-profiles behavior.

Ed3scomb3s:
Closing at the author request: a remaining all-profiles project-creation path reproduced after the initial fix. I will not resubmit until the complete behavior is verified.

PR #97337, opened 2026-08-28
fix(projects): bind-board writes the project link, not only the workdir mirror

rodrigogs:
fix(projects): bind-board writes the project link, not only the workdir mirror

`project bind-board` had two jobs and silently did one. It set the project's
`board_slug` and mirrored `default_workdir` onto the board, and never wrote the
board's own `project_id` — the field the inheritance path actually reads.

`kanban_db.write_board_metadata` documents what that field is for:

    ``project_id``: Optional first-class Project this board is scoped to. When
    set, new tasks inherit it (deterministic worktree + branch under the
    project's primary repo) and ``default_workdir`` mirrors the project's
    primary path.

and `create_task` implements exactly that. So the bind printed `Bound X -> board
Y` and the effect it exists for never arrived: every task kept landing on a
shared `dir` workspace, which is what `--project` exists to avoid.

MEASURED on one install before this fix: 11 of 11 boards had `project_id: null`
while their projects named them, and 249 tasks had piled onto shared working
trees — 119 of them on a single repo path. Two blocked each other with
"Paused to avoid concurrent edits in shared dir". The one board with worktree
tasks got them by passing `--workspace worktree` on every single create.

THE PRIMARY-PATH CHECK MAY NOT GATE THE LINK, which is the second half of the
same defect. The old code returned early when `primary_path` was empty, throwing
away the binding for want of the mirror's input. A project with no folder yet
still has an identity, and binding it is how a board starts inheriting one.

UNBIND NOW CLEARS WHAT BIND SET. Leaving `project_id` behind is how a board goes
on inheriting a project it is no longer bound to. `write_board_metadata` takes
`None` as "leave unchanged" and the empty string as "clear", so the clear has to
be deliberate.

Failures stay non-fatal, as before: the projects DB is the authority on the
binding and this mirror is a convenience on top of it.

Tests: five, in tests/hermes_cli/test_project_bind_board.py. The load-bearing one
goes through task creation — a task created on a bound board with no workspace
flags must land on a deterministic worktree under the project's repo — because
asserting the metadata alone would not have told us whether anything reads it.
Dropping `project_id` from the write fails four of the five.

One observation left for a separate change, not fixed here: `create_task`
resolves the board for that inheritance from its `board` argument, falling back
to the ambient current board — not from the connection it was handed. Creating a
task with a connection for board X while the active board is Y therefore inherits
Y's project. The `default_workdir` fallback a few lines below reads the board the
same way. Both are consistent with each other and predate this change.

### DIFF
--- a/hermes_cli/projects_cmd.py
+++ b/hermes_cli/projects_cmd.py
@@ -305,31 +305,82 @@ def _cmd_restore(args, conn, proj) -> int:
 
 @_with_project
 def _cmd_bind_board(args, conn, proj) -> int:
+    previous = (proj.board_slug or "").strip()
     pdb.update_project(conn, proj.id, board_slug=args.board)
     if args.board.strip():
         print(f"Bound {proj.slug} -> board {args.board}")
-        _sync_board_default_workdir(proj, args.board)
+        _sync_board_binding(proj, args.board)
     else:
         print(f"Unbound board from {proj.slug}")
+        # The board keeps a `project_id` of its own; leaving it set after an
+        # unbind is how a board goes on inheriting a project it is no longer
+        # bound to.
+        _clear_board_binding(previous)
     return 0
 
 
-def _sync_board_default_workdir(proj, board_slug: str) -> None:
-    """Best-effort: point the bound board's default_workdir at the primary repo.
+def _board_slug_or_none(board_slug: str):
+    """The normalized slug of an EXISTING board, or None. Shared by both writers."""
+    from hermes_cli import kanban_db as kb
 
-    Keeps kanban task worktrees anchored to the project's repo. Failures here
-    are non-fatal — the binding itself already succeeded.
+    slug = kb._normalize_board_slug(board_slug)
+    if not slug:
+        return None, kb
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related to #91191, which already covers reciprocal board project_id binding and additionally handles `project create --board`. This narrower bind-board repair overlaps that open broader PR.

rodrigogs:
Closing as a duplicate of #91191 (fangliquanflq, opened 2026-08-21 — a week before this PR).

Same root cause, same fix shape: `_cmd_bind_board` wrote `default_workdir` but never the board's own `project_id`, so tasks on a bound board never inherited the project. #91191 refactors `_sync_board_default_workdir` → `_sync_board_binding(bound=...)` to write `project_id` + `default_workdir` on bind and clear the reciprocal link on unbind — and additionally covers the `project create --board` path and guards the clear with a project_id match check, which this PR does not.

No unique delta here beyond test placement. Keeping one lane is cleaner for the maintainer pick; #91191 is the earlier, broader one.

PR #97344, opened 2026-08-28
fix(kanban): `kanban block` requires a reason like `kanban_block`

rodrigogs:
fix(kanban): `kanban block` requires a reason, like `kanban_block` always has

A blocked task is a request for a human decision, and a request with no sentence
is one nobody can act on: the next reader reconstructs the cause from run logs, or
unblocks blind and watches it re-block.

`kanban_block` — the tool workers call — has refused an empty reason from the
start: "reason is required — explain what input you need" (tools/kanban_tools.py,
`_handle_block`). The CLI accepted one, so whether the requirement applied
depended on which surface you happened to use for the same operation.

MEASURED on one install, over 141 `blocked` events: 126 carried a reason, 15 did
not, and 13 of those 15 came from outside any run — this path. So the worker-side
rule is working and this is the remaining hole, not the main one. The argument for
closing it is consistency rather than volume: two surfaces onto one operation
should not disagree about what the operation needs.

The refusal is total — exit 2 and the task is untouched. A task left blocked with
no reason is exactly the state the check exists to prevent, so half-applying it
would be worse than not checking.

Whitespace is not a reason: `kanban block t_x "   "` is refused too. The reason is
already stored twice on purpose and both keep working — the event payload, which a
report reads, and a `BLOCKED: <reason>` comment, which a person reads on the card.

Tests: seven, in tests/hermes_cli/test_kanban_block_requires_reason.py, including
the parametrised whitespace cases and one that asserts the worker tool's own
requirement is still the precedent this mirrors — if that message moves, this test
is the one that should notice. Removing the refusal fails five of the seven.

Pre-existing failures unchanged: `pytest tests/hermes_cli/ -k kanban` reports 15
failures on a clean upstream/main worktree and the same 15 with this change
applied (294 passed there, 301 here — the difference is this file's seven).

### DIFF
--- a/hermes_cli/kanban.py
+++ b/hermes_cli/kanban.py
@@ -907,6 +907,23 @@ def _commented(conn, reason: Optional[str], author, prefix: str, op):
 
 def _cmd_block(args: argparse.Namespace) -> int:
     reason = _joined_words(args.reason)
+    # A blocked task is a request for a human decision, and a request with no
+    # sentence is one nobody can act on: the next reader has to reconstruct the
+    # cause from run logs, or unblock blind and watch it re-block.
+    # ``kanban_block`` (tools/kanban_tools.py) has always refused an empty reason
+    # with "reason is required — explain what input you need". The same operation
+    # through the CLI accepted one, so the requirement depended on which surface
+    # you happened to use. Measured on one install: 126 of 141 block events
+    # carried a reason and 13 of the 15 that did not came from outside a run —
+    # this path.
+    if not reason:
+        print(
+            "kanban block: a reason is required — say what input or decision "
+            "the task is waiting for (the same rule kanban_block enforces for "
+            "workers)",
+            file=sys.stderr,
+        )
+        return 2
     kind = getattr(args, "kind", None)
     author = _profile_author()
     ids = _bulk_ids(args)
--- a/hermes_cli/kanban_parser.py
+++ b/hermes_cli/kanban_parser.py
@@ -292,7 +292,9 @@ _SPECS = [
     ], help="Edit recovery fields on an already-completed task"),
     _cmd("block", [
         _TASK_ID,
-        _arg("reason", nargs="*", help="Reason (also appended as a comment)"),
+        _arg("reason", nargs="*",
+             help="Required. Why it is blocked / what it waits for (also appended "
+                  "as a comment)"),
         _bulk_ids("block"),
         _arg("--kind", choices=sorted(kb.VALID_BLOCK_KINDS),
              help="Typed block reason. 'dependency' waits in todo (auto-promoted when "
--- a/tests/hermes_cli/test_kanban_block_requires_reason.py
+++ b/tests/hermes_cli/test_kanban_block_requires_reason.py
@@ -1,0 +1,103 @@
+"""`kanban block` requires a reason, like `kanban_block` always has.
+
+A blocked task is a request for a human decision, and a request with no sentence
+is one nobody can act on. The worker tool (tools/kanban_tools.py `_handle_block`)
+has always refused an empty reason with "reason is required — explain what input
+you need"; the CLI accepted one, so the requirement depended on which surface you
+used.
+
+Measured on one install: 126 of 141 `blocked` events carried a reason, and 13 of
+the 15 that did not came from outside any run — this path.
+"""
+
+from __future__ import annotations
+
+import argparse
+
+import pytest
+
+from hermes_cli import kanban as kanban_cmd
+from hermes_cli import kanban_db as kb
+from hermes_cli import kanban_db_connect as kbc
+
+
+@pytest.fixture(autouse=True)
+def _isolate(tmp_path, monkeypatch):
+    monkeypatch.setenv("HERMES_HOME", str(tmp_path))
+    monkeypatch.setattr(kb, "_HERMES_HOME_OVERRIDE", str(tmp_path), raising=False)
+    yield
+
+
+def _task():
+    conn = kbc.connect()
+    try:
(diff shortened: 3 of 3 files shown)


Enough1122:

**Overall:** Aligns CLI `hermes kanban block` with worker `kanban_block` contract: a blocked task must carry a reason.

**What it does**
- `hermes_cli/kanban.py:build_parser` help for `reason` now states "Required. Why it is blocked / what it waits for (also appended as a comment)".
- `_cmd_block` computes `reason = " ".join(args.reason).strip()` and if empty prints `kanban block: a reason is required — say what input or decision the task is waiting for (the same rule kanban_block enforces for workers)` to stderr and returns `2`. Adds comment citing measured 126/141 blocks carried reason, 13/15 missing came from this CLI path.

**Non-blocking notes**
- Return code `2` matches argparse error convention — consistent. This is a breaking change for scripts calling `kanban block <id>` without reason; add release note.
- Bulk `--ids` mode reuses same reason — still requires one reason, which is correct.

rodrigogs:
On the one substantive note here — the return code `2` being a breaking change for scripts that call `kanban block <id>` with no reason — the short answer is that the CLI is catching up to a contract the docs already state, and this repo has no release-note fragment to add to.

`website/docs/user-guide/features/kanban.md` already shows the reason as part of the invocation in both places it documents the verb: the overview example (`hermes kanban block t_abc "need input" --ids t_def t_hij`) and the CLI reference (`hermes kanban block <id> "<reason>" [--ids <id>...]`). So a script written against the documented surface never hits the new error — only one that relied on the undocumented reason-less form does, and that form produced a block row with no explanation, which is the bug this PR exists to close (measured: 13 of 15 reason-less blocks came from this CLI path, against 126 of 141 blocks that carried a reason).

I did not add a release-note file because there is no such convention in this repo — no `CHANGELOG.md` and no fragment requirement in the contributing docs — so a note here would be an orphan artifact. The PR title carries the behavior change, which is what the release tooling reads.

@Enough1122 thanks — the remaining notes are non-blocking and I am not acting on them: `--ids` reusing one reason is the intended shape (a single block operation, one reason, one audit comment per task), and the argparse-style exit code is exactly what makes the failure visible to a caller that was silently writing unexplained blocks.

Moep90:
Same conclusion from a different angle, with one suggestion: the refusal may be worth moving one layer down.

I independently hit the unexplained-block problem and put the check in `kanban_db.block_task` rather than in the CLI handler, because all three callers funnel through it: the tool (`tools/kanban_tools.py`, which has always refused a blank reason), this CLI path, and the dashboard drag in `plugins/kanban/dashboard/plugin_api.py`. The dashboard is the interesting one — a drag carries no prose at all, so before the change it wrote a NULL reason through a path the CLI check never sees.

Worth noting for that case: enforcing in `block_task` means the dashboard must supply something, and "blocked from the dashboard (no reason given)" is a better record than NULL. It states what actually happened, which an empty reason does not.

Not asking you to grow this PR — the CLI refusal is the gap operators hit, and your test coverage on it is good. Just flagging that the same reasoning applies to the dashboard, if a maintainer would rather close the whole class in one place.

alt-glitch:

Related: #99326 tracks the broader persisted reasonlessness problem; this patch validates only the CLI entry point.

rodrigogs:
@Moep90 — thanks for the independent read and, especially, for checking the dashboard path before suggesting.

For the record, what you describe on the dashboard is live today: `plugins/kanban/dashboard/plugin_api.py` routes the drag through `kanban_db.block_task(conn, tid, reason=getattr(p, "block_reason", None))`, and a drag carries no prose, so `reason` arrives as `None` and the row takes a NULL reason through the exact funnel you point at — a path the CLI check in this PR never sees. The class-wide gap is real.

I'm deliberately keeping this PR at the CLI entry point, for two reasons, both from the review history here: the first round asked for exactly one behavior (the CLI refusal), and the caller-side check is where the operator gets their feedback — an argparse error naming the flag beats a mid-write exception from a transaction helper.

But your layer is the right place for the class-wide guarantee, and the shape I'd suggest if you or a maintainer want it there: keep `block_task(reason=None)` legal (the dispatcher's own auto-block path calls `kanban_block` with no prose and must keep working), and instead reject a `reason` that is empty-after-strip at the helper boundary — `reason=""` / `"   "` raising `ValueError` — while the dashboard drag supplies `"blocked from the dashboard (no reason given)"`, your better-record point verbatim. That threads the needle: NULL-from-drag stops existing, the dispatcher is untouched, and the CLI refusal here stays the first line of defense rather than becoming redundant.

I've added that suggestion to the thread for whoever picks the follow-up up — happy to review it, but I don't want to grow this PR past the scope its first review round agreed, for the reason you already anticipated.

Enough1122:
Verified against current `main`: `website/docs/user-guide/features/kanban.md` does document the reason as part of the invocation in both places — the overview example (`hermes kanban block t_abc "need input" --ids ...`, kanban.md:397) and the CLI reference (`hermes kanban block <id> "<reason>" [--ids <id>...]`, kanban.md:958) — and the repo has no `CHANGELOG.md` at the root, so the no-fragment-convention point holds too.

Accept the docs-contract framing: my "breaking change / add a release note" note was overcautious — retracting it. The remaining non-blocking notes are fine left as-is.

rodrigogs:
@Enough1122 thanks for double-checking both `kanban.md` anchors and for the retraction — with the docs-contract point settled, nothing is open on this PR at the CLI entry point. The `kanban_db.block_task` layer stays available as the class-wide follow-up (as sketched for @Moep90 above) if a maintainer wants to close the dashboard-NULL case in the same move.

Issue #90292, opened 2026-08-19
Linux desktop entry generated with non-runnable Exec; icon launch always fails


1mems:
# Bug: Linux desktop entry is generated with a non-runnable `Exec`, icon launch always fails

**Version:** hermes 0.17.0 (desktop) / Hermes Agent v0.20.4 (2026.8.18), install method `git`
**OS:** Ubuntu 24.04.4 LTS, GNOME (X11)
**Install:** shell installer, venv at `<install>/venv`, launcher at `~/.local/bin/hermes`

## Summary

`hermes desktop` writes a `hermes.desktop` entry whose `Exec` points at the repo
script `<install>/hermes`. That script's shebang is `#!/usr/bin/env python3`, so
launching it from the desktop icon runs it under **system** Python, which has none
of the venv dependencies. It dies immediately:

```
Traceback (most recent call last):
  File "/home/<user>/.hermes/hermes-agent/hermes", line 10, in <module>
    from hermes_cli.main import main
  File ".../hermes_cli/main.py", line 697, in <module>
    from hermes_cli.env_loader import load_hermes_dotenv
  File ".../hermes_cli/env_loader.py", line 12, in <module>
    from dotenv import load_dotenv
ModuleNotFoundError: No module named 'dotenv'
```

Because the entry sets `Terminal=false`, nothing is shown to the user. GNOME shows a
spinner until startup notification times out. The CLI works fine, so the app appears
"installed but unlaunchable from the menu".

## Root cause

`~/.local/bin/hermes` is a bash wrapper:

```bash
exec "<install>/venv/bin/python" "<install>/hermes" "$@"
```

So within the process, `sys.argv[0]` is `<install>/hermes` — the script — not the wrapper.

`hermes_cli/relaunch.py::resolve_hermes_bin()` returns `argv[0]` when it is absolute,
a regular file, and `os.access(..., os.X_OK)`. `<install>/hermes` satisfies all three
(mode 775). `hermes_cli/linux_desktop_entry.py::resolve_exec_command()` then writes
that path into `Exec=`.

The check establishes that the file is *executable*, not that it is *independently
runnable*. Its `X_OK` bit passes; its shebang resolves to an interpreter without the
dependencies. On Linux the existing `_is_python_script()` guard does not apply — it is
Windows-only and keys off a `.py`/`.pyc` extension, and this file has no extension.

This is precisely the failure `linux_desktop_entry.py`'s module docstring says the
[...]

PR #97346, opened 2026-08-28
fix(desktop): stop baking the venv's base interpreter into the Linux Exec= line

shilai-li:
## Symptom

On Linux, the Hermes launcher icon does nothing. `hermes desktop` from a terminal works fine. The journal shows:

```
hermes.desktop[26049]: Traceback (most recent call last):
hermes.desktop[26049]:   File "~/.hermes/hermes-agent/venv/bin/hermes", line 4, in <module>
hermes.desktop[26049]:     from hermes_cli.main import main
hermes.desktop[26049]: ModuleNotFoundError: No module named 'hermes_cli'
```

Nothing is visible to the user because the entry sets `Terminal=false`.

## Root cause

`_needs_interpreter()` decided whether a launcher's shebang escapes the venv by comparing it against `Path(sys.executable).resolve().parent`.

In every symlinked venv — `uv`, `virtualenv`, `python -m venv --symlinks` — `venv/bin/python3` is a symlink to the base interpreter, so `.resolve()` yields the *base* interpreter's directory, which no venv console-script shebang can ever contain. The test inverted: a correct venv shebang was classified as escaping the venv, and `Exec=` was rewritten to run the venv's own entry-point script under the base interpreter.

Observed on an affected machine (Ubuntu 26.04, uv-managed CPython 3.11):

```
Exec=~/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/bin/python3.11 ~/.hermes/hermes-agent/venv/bin/hermes desktop
```

That interpreter has none of the venv's `site-packages`, including the editable install's `.pth` that provides `hermes_cli` — hence the import error.

`install_desktop_entry()` runs on every `hermes desktop` launch and keys off `argv[0]`, so the breakage is sticky rather than transient: once any launch resolves `argv[0]` to the venv console script, the entry stays broken.

The same `.resolve()` also weakened the two branches that *legitimately* prefix an interpreter (the `/usr/bin/env python3` case from #90292, and the `-m hermes_cli.main` fallback). Both baked in the base interpreter, which sees none of the venv's third-party dependencies — so the original #90292 fix was itself degraded on any symlinked venv.

## Fix

Drop `.resolve()` in all three places. `sys.executable` is already absolute, and it is the venv interpreter we actually want.

## Tests

`test_exec_leaves_venv_shebang_scripts_alone` wrote its fixture shebang from `Path(sys.executable).resolve()` — the one shape the buggy check happened to get right — which is why this was never caught. It now uses the real shape.

Added `test_exec_leaves_symlinked_venv_python_shebang_alone`, which builds a venv with a symlinked python and asserts the base interpreter never reaches `Exec=`. Reverting the source change makes it fail with exactly the broken line seen in the wild:

```
- .../venv/bin/hermes desktop
+ .../base/bin/python3.11 .../venv/bin/hermes desktop
```

`tests/hermes_cli/test_linux_desktop_entry.py`: 16 passed, 2 skipped.
`tests/hermes_cli/test_gui_command.py`, `test_gui_uninstall.py`: 47 passed, 9 skipped.

### DIFF
--- a/hermes_cli/linux_desktop_entry.py
+++ b/hermes_cli/linux_desktop_entry.py
@@ -78,11 +78,11 @@ def resolve_exec_command() -> str:
             # third-party import (#90292) — silently, since Terminal=false.
             # sys.executable is the interpreter actually running Hermes (the
             # venv one), so prefix it explicitly.
-            argv = [str(Path(sys.executable).resolve()), str(resolved), "desktop"]
+            argv = [sys.executable, str(resolved), "desktop"]
         else:
             argv = [str(resolved), "desktop"]
     else:
-        argv = [str(Path(sys.executable).resolve()), "-m", "hermes_cli.main", "desktop"]
+        argv = [sys.executable, "-m", "hermes_cli.main", "desktop"]
     return " ".join(_quote_exec_arg(a) for a in argv)
 
 
@@ -106,7 +106,12 @@ def _needs_interpreter(bin_path: Path) -> bool:
     # A python shebang pointing INSIDE the running interpreter's environment
     # already resolves correctly; anything else (``/usr/bin/env python3``,
     # a system path) would escape the venv when spawned by the DE.
-    exe_dir = str(Path(sys.executable).resolve().parent)
+    # Compare against the UNRESOLVED sys.executable. In a venv,
+    # ``venv/bin/python3`` is a symlink to the base interpreter, so
+    # resolving it yields a directory no venv console-script shebang can
+    # ever contain — the test inverts, and the venv-less base interpreter
+    # gets baked into ``Exec=``, reintroducing #90292 by another route.
+    exe_dir = str(Path(sys.executable).parent)
     return exe_dir not in shebang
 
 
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Fixes #90292 recurrence where venv interpreter symlink resolution baked base interpreter into Linux `Exec=` line.

**What it does**
- `hermes_cli/linux_desktop_entry.py:resolve_exec_command` now uses `sys.executable` directly (not `Path(...).resolve()`) when prefixing interpreter for non-executable `hermes` binary.
- `_needs_interpreter` now compares `str(Path(sys.executable).parent)` (unresolved) vs shebang instead of resolved parent. In a venv `venv/bin/python3` is symlink to base interpreter; resolved path never appears in shebang so check inverted and incorrectly injected base interpreter. Using unresolved parent restores correct test.
- `tests/hermes_cli/test_linux_desktop_entry.py` updated expectations.

**Non-blocking notes**
- `Exec=` will now contain symlink path (`venv/bin/python3`) — desired and matches other venv references. Ensure no downstream code assumes resolved path for Exec comparison.
- `_quote_exec_arg` still applied — correct for spaces.

No blocking issues.

shilai-li:
Superseded by #99492, which has now landed on main and covers the same symlinked-venv Exec= failure mode. Closing this PR as no longer needed.

PR #96891, opened 2026-08-28
feat(gateway): compact over-threshold sessions when a turn settles

SmelterLabs:
## What does this PR do?

An over-threshold TUI gateway session only compacts at the **next turn's preflight**. From the user's chair the session simply stops after a long turn; nothing happens while it sits idle; and when they finally send another message, that message blocks on a full summarization before any work resumes. The idle window right after the turn — the one moment compaction is free — goes unused.

This starts compaction at the **turn-finish boundary** instead. When a turn settles with no follow-up queued and the session is over threshold, a background daemon thread compacts while the session is idle, so the next prompt starts on an already-compacted transcript.

This is the background-timing delta that #32446 and #38956 were closed without. Both were superseded by #69360's `idle_compact_after_seconds`, which compacts long-idle sessions **at turn start** — it moves *which* turn pays, but a turn still pays. Both close-outs invited a fresh PR against current contracts if the latency proved painful in practice ("the right vehicle is a fresh issue/PR against the current contracts", #38956; "If background-timing proves worth it, a fresh issue against the current async session flow is the right vehicle", #32446). It did, so here it is, rebuilt against the contracts those reviews named.

Reported in practice as: a long session appears to stop, and "Compacting context" only begins after the user types something to ask what happened.

## Related Issue

Refs #35467 (progressive background pre-compaction — this implements the post-turn trigger half of that proposal, without the candidate-summary cache).
Refs #32446, #38956 (prior attempts, closed as superseded; this is the invited fresh vehicle).

## Type of Change

- [x] ✨ New feature (non-breaking change that adds functionality)

## Changes Made

`tui_gateway/server.py`

- `_maybe_post_turn_compress(sid, session)` — new. Runs the same automatic gate as preflight and, when it passes, compacts on a daemon thread.
- Called once at the end of the turn-finish path in `_run_prompt_submit`'s `run()`, after the queued-prompt drain, the turn-boundary continuation, the goal-continuation, and the completion-notification drain — so it only fires when nothing else is going to take the session.
- `_POST_TURN_COMPRESS_FAILURE_COOLDOWN_S = 300.0` — new module constant.

How it composes with the current contracts:

- **Automatic-compaction gate** — reuses `ContextCompressor.should_compress`, so the threshold, the summary-failure cooldown and the anti-thrash breaker all apply unchanged. Token estimate comes from `estimate_request_tokens_rough` over the session history with the agent's cached system prompt and tool schemas, matching preflight's accounting.
[...]

PR #93241, opened 2026-08-23
fix(compression): bound large-session compaction

Alish3r:
## Summary

- classify compression waits as idle timeouts versus absolute total-ceiling exhaustion and propagate the cause through callbacks, telemetry, cooldowns, and gateway hygiene
- cancel fence-owned auxiliary compression work without closing shared clients or allowing late transcript commits
- decide the Codex GPT-5.4/5.5/5.6 85% autoraise from the resolved effective context window instead of model/provider identity alone
- recompute thresholds from the configured base on model/provider/context switches while preserving explicit per-model thresholds and absolute token caps
- prevent lean-mode compaction from launching multi-chunk sequential auxiliary digest calls after the main summary has completed

## Why

Large Codex sessions had three related failure modes:

1. progressing summary streams could exhaust the absolute ceiling but be reported like idle/no-output timeouts;
2. a base Codex model resolving to a 900K effective window could still inherit the small-window 85% trigger, delaying local compression to roughly 765K instead of the configured 50%;
3. lean mode could perform up to 28 optional sequential digest calls after the main checkpoint, exhausting the host ceiling before commit.

The changes preserve transcript fencing, native Responses compaction, plugin context-engine ownership, gateway hygiene's separate policy, and the existing small-window floor.

## Verification

- broad compression/compaction/hygiene/threshold surface: 953 tests passed across 106 files
- current-main gateway compaction projection: 11 tests passed in a disposable CI-aligned environment
- Ruff: passed (one pre-existing invalid-noqa warning in `run_agent.py`)
- Python compilation: passed
- `git diff --check origin/main..HEAD`: passed
- no paid-provider call was needed for PR verification

## Safety

- no session-format, schema, dependency, or migration changes
- no historical session data is included or modified
- gateway hygiene keeps its independent 85% safety-net threshold
- native Responses compaction remains opt-in and clamps against the corrected local trigger

PR #97348, opened 2026-08-28
fix(compaction): stop idle compaction re-summarising a transcript that has not grown

JoaoMarcos44:
## Summary

Idle compaction (`compression.idle_compact_after_seconds`) skipped work only when the context was at or below `threshold_tokens × summary_target_ratio`. That is a **theoretical** post-compaction target — a real pass lands well above it, because the system prompt, the tool schemas and the protected head/tail are an incompressible floor.

So a session that compacted to well above that target stayed above it forever, and **every later idle resume re-ran a full summarisation over a transcript that had not grown.**

In the session reported in #97239:

| Time | Context | What ran | Blocked | `api_calls` |
|---|---|---|---|---|
| 17:01:10 | ~64,105 tok | idle compaction → **44,579 tok** | 651 s | 0 |
| 17:36:27 | ~44,579 tok | idle compaction **on the same transcript** | 256 s | 0 |

The second pass fired because `44,579 > 25,502` (the theoretical floor). Nothing had been added between them — **not one API call ran in either turn.** On a ~55 tok/s local llama.cpp route that is 15 minutes of blocked prompt reclaiming nothing.

`_should_idle_compact` now also honours `ContextCompressor.last_compression_rough_tokens` — what the previous pass on this session *actually produced*, already recorded by `compress_context` using the same `estimate_request_tokens_rough` shape the idle estimate uses. When it is known, the transcript must accumulate at least one `floor_tokens` worth of **new** content on top of it before another pass is worth its wall clock.

A raised floor is a **deferral, not an off switch**.

## Flow

```mermaid
graph TD
    A[🩸 Idle Resume] -->|gap >= idle_after_seconds| B{🔥 Failure Cooldown?}
    B -->|active| Z[⚔️ Skip Turn]
    B -->|clear| C{🕯️ Theoretical Floor<br/>threshold x target_ratio}
    C -->|at or below| Z
    C -->|above| D{💀 Last Pass Output Known?}
    D -->|no record| E[🔱 Compact Now<br/>original semantics]
    D -->|produced N tokens| F{🩸 Grown by one<br/>full floor above N?}
    F -->|not yet| Z
    F -->|yes| E
    E --> G[🔥 Summary Pass]
    G -->|records real output| H[(⚱️ last_compression_rough_tokens)]
    H -.->|honest floor for next resume| D
```

## Scope

This fixes **one** root cause behind #97239 — the repeated, unproductive compaction. It deliberately does **not** touch:

- the TUI status surface — owned by #97253
- the lean digest chain / absolute total ceiling — owned by #93241
- turn-settle compaction scheduling — owned by #96891

## Safety

- `last_compaction_tokens=0` (nothing compacted yet, or the counter cleared by a rebind/recalibration) keeps the original `tokens > floor_tokens` semantics **exactly**, so the first idle compaction of any session is unaffected.
- The call site type-pins the read: compressor doubles that expose a `Mock` there fall back to the original floor rather than raising it.
- Every other gate (opt-out, idle gap, failure cooldown) still wins over the raised floor — covered by a test.
- Reads existing state only. No config knob, no schema change, no new dependency, no persistence change.

## Testing

`tests/agent/test_idle_compaction.py` — 7 new predicate cases, including the exact #97239 numbers:

- unrecorded last compaction keeps the original floor
- skips when the transcript has not grown since the last pass
- fires once a full floor of new content accumulated
- boundary: does not fire at exactly the raised floor
- an effective pass still raises the floor
- other gates still win over the raised floor

`tests/agent/test_idle_compaction_lock_and_guards.py` — 3 new wiring cases proving `build_turn_context` reads the compressor value, defers, re-fires after growth, and ignores a non-int reading.

**Mutation-verified:** 5 of the new tests fail against pre-fix code.

**Regression sweep** — `pytest tests/agent/ -k "compaction or compression"`:

| | failed | passed |
|---|---|---|
| clean `origin/main` | 85 | 235 |
| this branch | 85 | **245** |

Identical failure set before and after; the 85 are a pre-existing local Python 3.14 breakage (`tools/daemon_pool.py` mirrors CPython 3.8–3.13 internals and references `ThreadPoolExecutor._initializer`, removed in 3.14) and are unrelated to this change. `tests/agent/test_turn_context.py`: 19 passed. Ruff clean, `py_compile` clean, `git diff --check` clean. No paid-provider call was needed.

### DIFF
--- a/agent/turn_context.py
+++ b/agent/turn_context.py
@@ -450,6 +450,7 @@ def _should_idle_compact(
     tokens: int,
     floor_tokens: int,
     cooldown_active: bool,
+    last_compaction_tokens: int = 0,
 ) -> bool:
     """Decide whether an idle-triggered compaction should run this turn.
 
@@ -465,6 +466,23 @@ def _should_idle_compact(
     *to*), so a small idle thread never pays for a summarisation that saves
     nothing, and it defers to an active compression-failure cooldown.
 
+    ``floor_tokens`` alone is a *theoretical* target (``threshold_tokens ×
+    summary_target_ratio``) that a real pass routinely misses: the system
+    prompt, the tool schemas and the protected head/tail are an
+    incompressible floor. A session that compacted to well above that target
+    therefore stays above it forever, so every later idle resume re-runs a
+    full summarisation over a transcript that has not grown — minutes of
+    silently blocked prompt on a slow route, reclaiming nothing (#97239).
+
+    ``last_compaction_tokens`` is what the previous pass on this session
+    actually produced (``ContextCompressor.last_compression_rough_tokens``,
+    the same ``estimate_request_tokens_rough`` shape as ``tokens``). When it
+    is known, require the transcript to have accumulated at least one
+    ``floor_tokens`` worth of *new* content on top of it before paying for
+    another pass. ``0`` — no compaction recorded yet, or the counter reset by
+    a rebind/recalibration — keeps the original floor semantics exactly, so
+    the first idle compaction of any session is unaffected.
+
     Pure predicate so the policy is unit-testable without a live agent.
     """
     if not enabled or idle_after_seconds <= 0:
@@ -473,7 +491,10 @@ def _should_idle_compact(
         return False
     if cooldown_active:
         return False
-    return tokens > floor_tokens
+    effective_floor = floor_tokens
+    if last_compaction_tokens > 0:
+        effective_floor = max(effective_floor, last_compaction_tokens + floor_tokens)
+    return tokens > effective_floor
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Stops idle compaction from re-summarising a transcript that compacted to an incompressible floor and will never dip below `floor_tokens` again (#97239).

**What it does**
- `agent/turn_context.py:_should_idle_compact` gains `last_compaction_tokens: int = 0` param and expanded docstring. When `last_compaction_tokens` is known (from `ContextCompressor.last_compression_rough_tokens`), requires `tokens > last_compaction_tokens + floor_tokens` (i.e., one floor worth of *new* content since last pass) before another pass. `0` keeps original `tokens > floor_tokens` semantics so first compaction unaffected and rebind/recalibration resets correctly.
- Call sites updated to thread last compaction tokens.

**Non-blocking notes**
- Default `0` preserves backward compat for callers not yet threading the value — safe.
- Ensure `last_compression_rough_tokens` is cleared on session rebind so growth check does not compare across sessions.

No functional issues.

kshitijk4poor:
Salvaged as #101649 — cherry-picked as-is with your authorship preserved, no changes needed. The 7 pure-predicate cases (including the exact #97239 numbers) and the 3 `build_turn_context` tests all pass on current main. Armed to merge on green.

Closing here so the merge shows against the salvage; credit stays yours in the git history.

Issue #93406, opened 2026-08-24
[Bug]: post-update fleet version check produces zero rows on a healthy resumed gateway — success path lacks the fail-closed check the restart-failure path already has


Halldrix:
## Bug Description

The post-update fleet version verification (#91277 Phase 1, `hermes_cli/update_receipt.py`) has an asymmetry: when the restart phase *fails*, the failure path already treats an empty fleet as incomplete via `_restart_phase_failure_is_incomplete` (`hermes_cli/update_cmd.py`, keyed on whether anything was running pre-update). When the restart phase *succeeds*, the success path applies **no** such cross-check: `collect_fleet_versions()` can return zero rows and `print_fleet_version_matrix()` early-returns (`if not fleet: return False`) — so an update that restarted a live gateway can complete with **zero fleet-check lines and exit 0**, and nothing in the output or receipt distinguishes "verified current" from "never verified".

Mechanically (verified against `main`): every failure path inside `collect_fleet_versions()` is swallowed via `logger.debug(...)` returning whatever accumulated so far, and the entire fleet call at the caller is wrapped in a catch-all `try/except Exception: logger.debug(...)`. An empty matrix and a swallowed exception are indistinguishable from a healthy fleet in both output and receipt.

**Relation to #93363:** that PR fixes the *opposite* failure mode of this check — a recycled PID reported falsely `STALE` (the false positive in #93258) — by swapping `_pid_exists` for `runtime_status_pid_is_live()` in the JSON fallback path. That fix is correct and orthogonal to this report: this one is about the **silent/false-negative** class, which remains open after that patch because it happens upstream of any row being produced.

## Steps to Reproduce

Real W11 box (venv install, single profile, Python 3.11.15). Gateway running via a Scheduled Task launched as `python -m hermes_cli.main gateway run` (no live `hermes.exe` shim, so the self-lock guard doesn't abort). Three runs of `python -m hermes_cli.main update --yes`, three main SHAs:

| Run | Commits | Pause→Resume | Fleet check block | Exit |
|---|---|---|---|---|
| 1 (`8db59d08f4`→`45fcaaa54a`) | 9 | ✓ | **absent** | 0 |
| 2 (`45fcaaa54a`→`175054c14b`, incl. divergent-reset) | 13 | ✓ | **absent** | 0 |
| 3 (`175054c14b`→`d962b21fa4`) | 3 | ✓ (+force-stop of 1 proc) | **absent** | 0 |

[...]

PR #97350, opened 2026-08-28
fix(update): scope fleet-probe runtime expectation to gateway-kind plan records

### LINKED ISSUE #97332: [Bug]: dashboard-only runtime makes successful `hermes update` exit 1 on empty gateway fleet check
## Bug description

`hermes update` exits 1 after successfully updating the code when a dashboard is running but no gateway is running.

The pre-update inventory includes dashboards and `hermes serve` processes in `UpdatePlan.runtimes`. `_fleet_probe_expected_runtimes()` currently treats any non-empty `pre_update_plan.runtimes` as evidence that the gateway fleet probe must return rows. However, `collect_fleet_versions()` reports gateway identities only. A dashboard-only plan therefore creates an impossible expectation: the updater waits for gateway rows that cannot exist, prints the incomplete-verification warning, records a partial result, and exits 1.

This is adjacent to #93406 and merged PR #93654, but it is the opposite boundary condition: their fail-closed correction correctly protects real gateways, while the plan-level signal is broader than the gateway-only verifier. The mismatch became reachable when merged PR #95576 later added `dashboard` and `serve` records to the same plan inventory. PR #95897 removed one non-row-capable Windows signal but retained the unfiltered plan-runtime check.

## Steps to reproduce

1. Use a git-installed Hermes checkout.
2. Run the Hermes dashboard and leave the gateway stopped.
3. Confirm the update plan contains only the dashboard:

```text
Update plan:
  Running services to restart (1):
    • dashboard [default] ... — manual-serve
```

4. Run `hermes update`.

The update completes and the dashboard restarts, then the updater prints:

```text
✓ Update complete!

⚠ Fleet version check returned no rows even though gateway runtimes were expected — verification incomplete.
```

The stock updater exits 1 even though the checkout is current and `hermes doctor` passes.

## Expected behavior

A dashboard-only or serve-only runtime should not require gateway-version rows. If no gateway-specific liveness signal fired, an empty gateway fleet snapshot should not turn a successful code update into exit 1.

The existing fail-closed behavior must remain unchanged when a gateway was running, restarted, killed, discovered in the pre-restart PID snapshot, or represented by a gateway runtime in the plan.

## Actual behavior

Any runtime in `pre_update_plan.runtimes`, including `kind="dashboard"`, makes `_fleet_probe_expected_runtimes()` return `True`. The post-update gateway probe returns no rows, so the update is classified as partial and exits 1.

## Root cause

Confirmed in installed upstream `baa344de` and still present on live `main` at `00bbfc6900`:

- `collect_runtime_inventory()` appends `RuntimeRecord(kind="dashboard", ...)` and `RuntimeRecord(kind="serve", ...)` entries to `UpdatePlan.runtimes`.
- `_fleet_probe_expected_runtimes()` checks only whether `pre_update_plan.runtimes` is non-empty.
- `collect_fleet_versions()` verifies gateways, not dashboard/serve runtimes.
- `test_incomplete_when_pre_update_plan_saw_runtimes` uses `_plan([object()])`, so it does not pin the required runtime-kind distinction and cannot catch the dashboard-only false positive.

## Suggested correction

Make the plan-derived expectation gateway-specific, for example by requiring at least one runtime whose `kind == "gateway"`, while preserving the existing gateway-specific signals from restart bookkeeping and the pre-restart PID snapshot.

Add regression coverage for both sides:

1. `UpdatePlan.runtimes = [RuntimeRecord(kind="dashboard", ...)]` with no gateway signals returns `False`.
2. A `kind="serve"`-only plan also returns `False` if the gateway verifier cannot produce rows for it.
3. `UpdatePlan.runtimes = [RuntimeRecord(kind="gateway", ...)]` still returns `True`.
4. Existing restarted/killed/pre-restart-gateway-PID cases remain fail-closed.

## Risk and scope

The change should be limited to deciding whether gateway rows are expected. It should not weaken stale/down gateway detection or remove dashboard/serve runtimes from the broader update plan and restart reconciliation.

## Environment

- Linux x86_64, kernel 6.8.0
- Hermes Agent v0.20.6, upstream `baa344de`
- Python 3.11.15
- Git install

## Plain English

The updater sees a dashboard, calls it evidence that a gateway must exist, and then fails because it cannot find that nonexistent gateway. Filter the expectation by runtime kind so real gateway verification remains strict without making dashboard-only updates look broken.

liuhao1024:
## What does this PR do?

`hermes update` exits 1 after a *successful* update when a dashboard (or `hermes serve`) is running but no gateway is. The pre-update plan inventory carries `kind="dashboard"` / `kind="serve"` records (#95576), and `_fleet_probe_expected_runtimes()` treated any non-empty `pre_update_plan.runtimes` as evidence that gateway fleet rows must exist. `collect_fleet_versions()` reports gateway identities only, so a dashboard-only plan created an impossible expectation: the probe waited for rows that cannot exist, printed the "verification incomplete" warning, recorded a partial result, and exited 1.

The fix keys the plan-derived expectation on `kind == "gateway"` records — the same row-capability rule the function already applies to the Windows resume token (#93406: "expected-runtimes must key only on signals that map to rows the probe can actually see"). A serve/dashboard backend never publishes the `gateway_state.json` row the probe would need, so it cannot ground a rows-expected verdict. All existing fail-closed signals (restart-phase bookkeeping, killed PIDs, pre-restart PID snapshot including the unreadable `None` case) are untouched, and a real gateway record in the plan still expects rows.

Not related to #78863 (Node-dependency repair at current HEAD) — that PR touches `hermes_cli/update_cmd.py` too but a different function, and does not touch the fleet probe.

## Related Issue

Fixes #97332

## Type of Change

## Changes Made

- `hermes_cli/update_cmd.py` — `_fleet_probe_expected_runtimes()`: the plan check now requires ≥1 runtime with `kind == "gateway"` instead of any non-empty `runtimes`; docstring documents the gateway-scoping rationale (#97332) alongside the existing token-exclusion precedent.
- `tests/hermes_cli/test_update_fleet_check_fail_closed.py` — replace the `object()` placeholder in `test_incomplete_when_pre_update_plan_saw_runtimes` with a real `RuntimeRecord(kind="gateway", ...)` so the test pins the runtime-kind distinction; add `test_dashboard_only_plan_does_not_expect_rows`, `test_serve_only_plan_does_not_expect_rows`, and `test_mixed_plan_keys_on_gateway_record`.
- `tests/hermes_cli/test_update_fleet_probe_resume_token.py` — same `object()` → `RuntimeRecord(kind="gateway", ...)` update for the plan-signal test.

## How to Test

1. `uv run pytest tests/hermes_cli/test_update_fleet_check_fail_closed.py tests/hermes_cli/test_update_fleet_probe_resume_token.py -v` — Observed result: **20 passed** (16 pre-existing incl. the updated ones + 4 new regression cases).
2. New negative cases pin the bug: a plan of `[RuntimeRecord(kind="dashboard", ...)]` or `[RuntimeRecord(kind="serve", ...)]` with no gateway-side signals returns `False` (pre-fix the non-empty check returned `True`, which produced exit 1 on a healthy dashboard-only update).
3. Fail-closed controls still hold: `kind="gateway"` plan record → `True`; restarted services / killed PIDs / non-empty or unreadable pre-restart PID snapshot → `True` (covered by the unchanged pre-existing tests in both files).
4. Broader local run `uv run pytest tests/hermes_cli/ -k "update" -q`: 754 passed, 48 failed — all 48 reproduce on a clean `upstream/main` checkout of this branch point (or pass when run in isolation), i.e. pre-existing local-environment failures / order coupling, none in the two files this PR touches.

### Code

## For New Skills

N/A — not a skill.

### DIFF
--- a/hermes_cli/update_cmd.py
+++ b/hermes_cli/update_cmd.py
@@ -10066,7 +10066,7 @@ def _fleet_probe_expected_runtimes(
     * ``pre_restart_pids`` non-empty, or ``None`` (pre-state unreadable —
       cannot prove nothing was running; same contract as
       ``_restart_phase_failure_is_incomplete``, #78574).
-    * the pre-update plan inventoried ≥1 runtime.
+    * the pre-update plan inventoried ≥1 runtime with ``kind == "gateway"``.
 
     ``windows_resume_token`` is deliberately EXCLUDED (#93406 residual). The
     pause/resume token is bookkeeping for ``_pause_windows_gateways_for_update``
@@ -10089,6 +10089,18 @@ def _fleet_probe_expected_runtimes(
     parameter stays in the signature so the call site keeps passing the token
     (cheap, explicit, and the docstring is where the exclusion is explained).
 
+    The plan-runtime check is gateway-scoped for the same reason (#97332):
+    the plan inventory also carries ``kind="serve"`` / ``kind="dashboard"``
+    records (#95576), and ``collect_fleet_versions()`` verifies gateway
+    identities only — a serve/dashboard backend never publishes the
+    ``gateway_state.json`` row the probe would need. Counting any non-empty
+    plan therefore made a dashboard-only update (gateway never started)
+    wait out the probe for rows that cannot exist, print the
+    incomplete-verification warning, and exit 1 after a successful update.
+    Only a ``kind == "gateway"`` record is evidence that gateway rows were
+    expected; the restarted/killed/pre-restart-PID signals above still fail
+    closed unchanged.
+
     The same condition gates the 2.0s settle sleep: a freshly restarted
     gateway needs the settle window to rewrite ``gateway_state.json``.
 
@@ -10103,7 +10115,15 @@ def _fleet_probe_expected_runtimes(
         return True
     try:
         if pre_update_plan is not None and pre_update_plan.runtimes:
-            return True
+            # Gateway-scoped on purpose (#97332): serve/dashboard plan
+            # records have no collect_fleet_versions() row, so only a
+            # kind == "gateway" runtime is evidence that gateway rows were
+            # expected.
+            if any(
+                getattr(runtime, "kind", None) == "gateway"
+                for runtime in pre_update_plan.runtimes
+            ):
+                return True
     except Exception:
         pass
     return False
--- a/tests/hermes_cli/test_update_fleet_check_fail_closed.py
+++ b/tests/hermes_cli/test_update_fleet_check_fail_closed.py
@@ -12,9 +12,11 @@ which never fires on Windows: ``_pause_windows_gateways_for_update`` /
 hoists the "should the probe have produced rows?" decision into
 ``_fleet_probe_expected_runtimes`` and keys it on the ROW-CAPABLE pre-update
 liveness signals: restart-phase bookkeeping, the pre-restart PID snapshot,
-and the pre-update plan inventory.  The Windows pause/resume token is
-deliberately NOT a signal — it is bookkeeping, not a runtime inventory, and
-its entries have no corresponding ``collect_fleet_versions()`` rows (see
+and the gateway-kind records in the pre-update plan inventory (the plan's
+serve/dashboard records are row-incapable for this probe, #97332).  The
+Windows pause/resume token is deliberately NOT a signal — it is bookkeeping,
+not a runtime inventory, and its entries have no corresponding
+``collect_fleet_versions()`` rows (see
 ``test_update_fleet_probe_resume_token.py``).  The same condition gates the
 2.0s settle sleep.
 """
@@ -25,6 +27,7 @@ import inspect
 import types
 
 from hermes_cli.main import _fleet_probe_expected_runtimes
+from hermes_cli.update_inventory import RuntimeRecord
 
 
 def _plan(runtimes):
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** Scopes fleet-probe expectation to `kind == "gateway"` runtimes — fixes false incomplete-verification warning on dashboard/serve-only updates (#97332, #95576).

**What it does**
- `hermes_cli/update_cmd.py:_fleet_probe_expected_runtimes` docstring now explains `windows_resume_token` exclusion (#93406) and gateway-scoping rationale: plan inventory carries `kind="serve"`/`"dashboard"` records but `collect_fleet_versions()` verifies gateway identities only, so counting any non-empty plan made dashboard-only update wait for rows that cannot exist and exit 1 after success. Logic changes `if pre_update_plan.runtimes: return True` to check `any(r.kind == "gateway" for r in rurimes)`. Same gate controls 2.0s settle sleep. Restarted/killed/pre-restart-PID signals still fail closed.
- `tests/hermes_cli/test_update_fleet_check_fail_closed.py` + `test_update_fleet_probe_resume_token.py` updated.

**Non-blocking notes**
- Historical plans without `kind` field will be treated as non-gateway (no probe) — fail-open for probe but still covered by other signals; consider treating missing kind as gateway for backward compat if needed.
- `windows_resume_token` stays in signature for explicit call site but excluded from logic — correct and documented.

liuhao1024:
Thanks for the review. On the historical-plan note: plans without `kind` predate the fleet-probe contract, and those updates still fail closed through the restart/killed/PID signals — I kept the strict gateway-only reading rather than widening the probe net on a shape we can't distinguish from a serve-only plan. Happy to flip missing-kind to gateway if a stale-plan repro ever shows up.

teknium1:
Landed in #103478 (f58fcc8118) — your commit cherry-picked with authorship, applied at the function's current home (`hermes_cli/update_cmd_fleet.py`); the three plan-kind tests were folded into one invariant test. Fixes #97332. Thanks!

PR #97351, opened 2026-08-28
feat(kanban): a retry after a no-verdict exit is told what the last run left behind

rodrigogs:
feat(kanban): a retry after a no-verdict exit is told what the last run left behind

`detect_crashed_workers` (now in `hermes_cli/kanban_db_dispatch.py`) already handles
the clean-exit-without-verdict case well: it records a protocol violation, and its
corrective sentence reaches the retry worker through `build_worker_context` — "if the
prior run already did the work, verify it and report the result via kanban_complete".

Verifying it began with REDISCOVERING it. Nothing handed the retry what the
previous run had changed, so every attempt re-derived the same state from scratch
before it could act on the guidance.

MEASURED on one install: 117 runs ended this way across 49 tasks — 5.0 runs per
task — and 91% of those tasks did eventually complete, which is the point. The
guidance works; the cost is the rediscovery. Four tasks never recovered at all,
one of them sitting blocked for five weeks with its work already committed.

WHAT THIS ADDS. When the most recent closed attempt was a no-verdict exit and the
task has a `dir` or `worktree` workspace that is a git repo, the worker prompt now
names what is there: commits whose committer date falls inside that run's window,
and a count of uncommitted files. Committer date rather than author date on
purpose — a commit authored earlier but committed by that run still landed during
it, and that is the case worth catching.

The section is emitted from `_ctx_prior_attempts` in `hermes_cli/kanban_db.py`,
right after the shown-run tail, so it sits beside the attempt whose evidence it is.
The violation is recognised the same way `_protocol_violation_streak` recognises it
(the durable `protocol_violation` run-metadata marker, with the error text as
fallback for runs recorded before the marker existed).

"NOTHING LANDED" IS ALSO AN ANSWER, and it is why the empty case is not silent:
telling the retry the workspace is untouched saves it from hunting for work that
was never there.

BOUNDARIES, each with a test:

* Only the no-verdict case. A run that reported its own failure already said what
  happened; a workspace dump on top of that is noise.
* Only `dir` / `worktree`. A `scratch` task is not a repo checkout, and probing a
  path it happens to carry would be a guess about someone else's tree.
* A workspace that is not a repo, or a git call that fails or times out, produces
  nothing at all. A prompt that cannot be built is worse than a prompt without
  this section, and the guidance it sits beside already works without it.
* Read-only, capped at five commits, ten-second timeout.

Everything it reports is labelled unverified, because the run that produced it
never reported it. The retry has to confirm before claiming it — inheriting an
unreported claim is the failure this whole path exists to catch.

Tests: seven, in tests/hermes_cli/test_kanban_no_verdict_evidence.py. Three
mutations were run against them: not calling the evidence at all, dropping the
`--since` window, and dropping the workspace-kind guard.

The third one SURVIVED the first time, and the reason is worth recording: the
scratch test used a task with no workspace path, so it passed on the empty-path
check and said nothing about the kind. It now points a scratch task at a real repo
with a real commit, which makes the kind guard the only thing standing between this
feature and a tree it has no claim to read — and then the mutation fails.

The repo fixture backdates its seed commit for the same class of reason: with the
seed at "now" it fell inside the run window and the empty-workspace assertions were
passing on the fixture's own noise.

Rebased onto current main. The test opens its connection through
`hermes_cli.kanban_db_connect.connect` (the `kanban_db.connect` alias is a
plugin-compat pointer that `scripts/check_compat_pointers.py` rejects for in-tree
code). Locally green: the seven new tests, `test_kanban_db.py`,
`test_kanban_core_functionality.py`, `test_kanban_blocked_sticky.py`,
`tests/plugins/test_kanban_attachments.py`, ruff, and the compat-pointer check.

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -307,6 +307,14 @@ def _resolve_rate_limit_cooldown_seconds() -> int:
 # build_worker_context() caps, sized for a ~100k-char prompt with headroom.
 _CTX_MAX_PRIOR_ATTEMPTS = 10      # most recent N prior runs shown in full
 _CTX_MAX_COMMENTS       = 30      # most recent N comments shown in full
+# A clean-exit protocol violation means the run did its work and skipped the
+# paperwork — the guidance already tells the retry to "verify it and report the
+# result via kanban_complete". Verifying it started with REDISCOVERING it: nothing
+# handed the retry what the previous run had changed, so each attempt re-derived
+# the same state. Measured on one install: 117 such runs across 49 tasks, 5.0 runs
+# per task, and 4 tasks that never recovered at all.
+_CTX_EVIDENCE_MAX_COMMITS = 5
+_CTX_EVIDENCE_TIMEOUT_S = 10
 _CTX_MAX_FIELD_BYTES    = 4 * 1024   # per summary/error/metadata/result
 _CTX_MAX_BODY_BYTES     = 8 * 1024   # per task.body (opening post)
 _CTX_MAX_COMMENT_BYTES  = 2 * 1024   # per comment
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Surfaces what a clean-exit-without-verdict run left behind so the retry does not have to rediscover it (measured: 117 such runs across 49 tasks, 5.0 runs/task, 4 never recovered).

**What it does**
- `hermes_cli/kanban_db.py` adds `_CTX_EVIDENCE_MAX_COMMITS=5`, `_CTX_EVIDENCE_TIMEOUT_S=10`, `_run_was_protocol_violation(run)` (checks `metadata.protocol_violation` or `"protocol violation"` in error), `_protocol_violation_evidence(task, run)` (git log `--since` committer date for commits landed during that run, read-only, time-bounded, silent on any failure), and threads evidence into worker context builder. Guidance already tells retry to verify and report via `kanban_complete`; this gives it the commits to verify.
- `tests/hermes_cli/test_kanban_no_verdict_evidence.py` covers both marker paths.

**Non-blocking notes**
- `--since` on committer date correctly captures commits authored earlier but committed at end — the case worth catching. Timeout 10s + max 5 commits bounds cost; silent failure keeps prompt build from failing — correct tradeoff.
- Consider including commit subject lines in addition to hashes for readability (already does if test shows).

No blocking issues.

teknium1:
Adjacent landing FYI: #112618 (2cfb655d52) puts the worker's own last output onto `last_failure_error` and the reap event, and `build_worker_context` already surfaces `last_failure_error` to the retry worker — so a retry after a no-verdict exit now sees what the previous run *said*. What it does not carry is what the run *left behind* (files touched, branch state), which is this PR's actual scope; leaving it open on that basis.

rodrigogs:
Ack — thanks for checking the boundary before commenting. Agreed on both halves: #112618 covering what the previous run *said* leaves this PR's actual scope (what the run *left behind*: branch state, touched files) intact as the open delta, which is exactly what the retry worker still has to discover by inspection today. Leaving it open on that basis is right. Not rebasing or pushing to this branch; it stays `mergeable` and ready for review as-is.

PR #85633, opened 2026-08-13
fix(scripts): expand check-windows-footguns --all coverage to full repository

Lesnak1:
## What does this PR do?

`scripts/check-windows-footguns.py --all` previously hardcoded only 8 subdirectories (`hermes_cli`, `gateway`, `tools`, `cron`, `agent`, `plugins`, `scripts`, `acp_adapter`), leaving ~100+ files and packages (including top-level modules like `cli.py`, `run_agent.py`, `hermes_state.py`, and directories `evals/`, `website/`, `providers/`, `skills/`) untracked.

This PR expands `--all` to dynamically include all top-level `*.py` files and missing repository directories while fixing an uncovered `Path.read_text()` missing `encoding="utf-8"` in `evals/readtool/report.py`.

## Related Issue

Fixes #85639
 (Discovered via CI full-repo test audit)

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `scripts/check-windows-footguns.py`:
  - Added `evals`, `website`, `providers`, `skills`, and top-level `*.py` files to `--all` scan roots.
- `evals/readtool/report.py`:
  - Added explicit `encoding="utf-8"` to `rep_file.read_text()`.

## How to Test

1. Run full-repo footgun scan:
   ```bash
   python scripts/check-windows-footguns.py --all
   
pytest tests/scripts/test_windows_footguns_full_repo_scan.py -v

PR #84323, opened 2026-08-12
fix(tui_gateway): make host PID liveness checks Windows-safe

ryantuc:
## What does this PR do?

`tui_gateway.host_supervisor._pid_alive()` used the POSIX `os.kill(pid, 0)` liveness idiom. On Windows CPython, signal 0 aliases `CTRL_C_EVENT` and is routed through `GenerateConsoleCtrlEvent`, so a nominal PID probe can broadcast Ctrl+C to processes sharing the target's console group.

This routes the probe through the existing canonical `gateway.status._pid_exists()` helper instead of creating another liveness policy. That helper prefers the core `psutil` dependency, treats zombies as dead, and retains the project's platform-specific stdlib fallbacks. If the helper unexpectedly raises, `_pid_alive()` conservatively assumes the PID is alive because uncertainty is not proof of death and a false-dead result can trigger duplicate/destructive recovery behavior.

The timeout path also resolves the force-termination signal with `getattr(signal, "SIGKILL", signal.SIGTERM)`. At this site the old bare `signal.SIGKILL` reference was inside `_terminate_pid()`, so on Windows it raised `AttributeError` only when that timeout path executed, not at import time.

## Related Issue

Related to #57145, but not a duplicate: #57145 reports the same Windows `os.kill(pid, 0)` mechanism in `scripts/run_tests_parallel.py`; this PR fixes `tui_gateway/host_supervisor.py`.

Related to #68031, but not a duplicate: #68031 applies the canonical-helper approach to `gateway/delivery_ledger.py`; this PR addresses a different remaining call site in the compute-host supervisor.

Neither item is closed by this PR, so this deliberately does not say `Fixes #57145`.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Route `tui_gateway/host_supervisor.py::_pid_alive` through `gateway.status._pid_exists` with a function-local import.
- Treat unexpected helper failures as uncertain/alive rather than falsely dead.
- Fall back to `SIGTERM` when `signal.SIGKILL` is unavailable on the timeout path.
[...]

PR #97352, opened 2026-08-28
fix(scripts): --all skipped tui_gateway and every package outside its 8 hardcoded roots

### LINKED ISSUE #97019: [Bug]: host_supervisor uses os.kill(pid, 0) for liveness, unsafe on Windows

BkashJEE:
## What does this PR do?

`scripts/check-windows-footguns.py --all` advertises a full-repository scan, and it is what the blocking **"Windows footguns (blocking)"** lint job runs (`.github/workflows/lint.yml:180`). It actually walked a hand-maintained list of eight directories — `hermes_cli`, `gateway`, `tools`, `cron`, `agent`, `plugins`, `scripts`, `acp_adapter` — and silently ignored everything else in the repo.

So the sweep reported a clean tree for code it had never opened:

```console
$ python scripts/check-windows-footguns.py --all
✓ No Windows footguns found (1021 file(s) scanned).

$ python scripts/check-windows-footguns.py tui_gateway/host_supervisor.py
✗ 2 Windows footgun(s) found across 1 file(s) scanned.
tui_gateway/host_supervisor.py:92: [os.kill(pid, 0)]
tui_gateway/host_supervisor.py:547: [bare signal.SIGKILL]
```

`tests/scripts/test_windows_footguns_full_repo_scan.py` exists precisely to stop a bare `os.killpg`/`signal.SIGKILL` regression reaching main, and its docstring records that one already had. It passed throughout — it shells out to `--all`, so it inherited the same blind spot.

`--all` now walks the repository. Coverage goes **1021 → 1168 files**, picking up `tui_gateway/`, `evals/`, `providers/`, `skills/` and the top-level modules (`cli.py`) that a directory-only list could never see.

`tests/` is skipped by the sweep alone. It is the larger half of the repo and currently carries ~2.3k matches, nearly all bare `read_text()`, which would bury the production hits this job exists to surface. Explicit paths, `--diff` and staged mode still scan it, so a footgun added to a test is still caught at review time; that backlog is its own campaign.

### Why this approach

Walking the repo and subtracting a small declared skip list means anything added at the repo root is covered the day it lands, rather than the day someone remembers to extend a list. The skip set is a named constant with a comment, and a test asserts it stays narrow.

## Related Issue

Fixes #97019

## Type of Change

## Changes Made

**The coverage gap**

- `scripts/check-windows-footguns.py` — added `full_scan_roots()`; `--all` uses it instead of eight hardcoded paths. Added `FULL_SCAN_SKIP_DIRS = {"tests"}` (sweep-only, documented). Updated `--all` help text, which claimed a full-repo scan.

**The seven findings that surfaced** — fixed here so the blocking job stays green:

- `tui_gateway/host_supervisor.py:92` — `_pid_alive` used `os.kill(pid, 0)`. On Windows CPython maps `sig=0` onto `CTRL_C_EVENT` and routes it through `GenerateConsoleCtrlEvent(0, pid)`, Ctrl+C-ing the entire console process group containing the target ([bpo-14484](https://bugs.python.org/issue14484)) — the case `CONTRIBUTING.md`'s first critical rule forbids. `reconcile_startup_orphan` calls it as a **read-only probe** before deciding whether to touch anything, so on Windows the probe was killing the process it was only meant to ask about, plus unrelated siblings sharing its console. Now delegates to `gateway.status._pid_exists`, matching `cli.py`, `cron/executions.py`, `gateway/delivery_ledger.py` and `gateway/lifecycle_ledger.py`.
- `tui_gateway/host_supervisor.py:547` — `_terminate_pid`'s escalation referenced `signal.SIGKILL`, which does not exist on Windows. The `AttributeError` was swallowed by the surrounding `except` and logged at debug, so **the force-kill silently never ran**. Now `getattr(signal, "SIGKILL", signal.SIGTERM)`; on Windows `os.kill` reaches `TerminateProcess`, which is the force-kill this branch wants. (Not reported in #97019 — it surfaced from the same scan.)
- `evals/readtool/report.py`, `evals/session_search_schema/runner.py`, `apps/desktop/scripts/perf/gateway_attach_bench.py` — five bare `read_text()`/`write_text()` calls now pass `encoding="utf-8"`.

**Tests**

- `tests/scripts/test_windows_footguns_full_repo_scan.py` — three tests: the walk reaches packages the old list omitted; the skip list stays narrow and declared; naming `tests/` explicitly still scans it.

## How to Test

1. Confirm the gap on `main`:
   ```console
   $ python scripts/check-windows-footguns.py --all
   ✓ No Windows footguns found (1021 file(s) scanned).
   $ python scripts/check-windows-footguns.py tui_gateway/host_supervisor.py
   ✗ 2 Windows footgun(s) found across 1 file(s) scanned.
   ```
2. On this branch, the sweep sees them and is clean after the fixes:
   ```console
   $ python scripts/check-windows-footguns.py --all
   ✓ No Windows footguns found (1168 file(s) scanned).
   ```
3. `pytest tests/scripts/test_windows_footguns_full_repo_scan.py tests/scripts/test_footgun_subprocess_encoding.py -q` → 13 passed.
4. `ruff check .` → passes.
5. `_pid_alive` behaviour, verified on Windows 11: `True` for the current process and a live child, `False` after that child is killed, `False` for `0`, a negative pid and an unused high pid.

I deliberately did **not** script a live reproduction of the `os.kill(pid, 0)` kill on my own machine — `GenerateConsoleCtrlEvent` would have signalled unrelated processes in my session. The mechanism is documented in `gateway/status._pid_exists`'s own docstring and in bpo-14484.

### Note on unrelated test failures

`tests/tui_gateway/test_compute_host.py::test_compute_host_line_json_seed_turn_interrupt` and `tests/tui_gateway/test_compute_host_phase1.py::test_append_log_record_single_write_lines` fail for me on Windows on **clean `origin/main`** as well as on this branch. They are pre-existing and unrelated to this change — I have not touched them.

### Code

- [x] I've run the relevant tests and they pass (see note on two pre-existing unrelated failures above)

- [x] I've updated relevant documentation (`--all` help text and in-code comments) — no user-facing docs reference this script
- [x] N/A — no config keys added or changed
- [x] Cross-platform impact considered: this is a Windows correctness fix; `_pid_exists` and `getattr(signal, ...)` are no-ops behaviourally on POSIX

### DIFF
--- a/apps/desktop/scripts/perf/gateway_attach_bench.py
+++ b/apps/desktop/scripts/perf/gateway_attach_bench.py
@@ -114,7 +114,7 @@ def main() -> int:
     content_b64 = base64.b64encode(png_bytes(args.kb)).decode("ascii")
 
     scratch = home / "scratch.txt"
-    scratch.write_text("hello from the bench\n")
+    scratch.write_text("hello from the bench\n", encoding="utf-8")
 
     image_on_disk = home / "on-disk.png"
     image_on_disk.write_bytes(png_bytes(args.kb))
--- a/evals/readtool/report.py
+++ b/evals/readtool/report.py
@@ -28,7 +28,7 @@ def load_label(label: str, model_filter: str | None) -> dict:
         if model_filter and model_dir.name != model_filter:
             continue
         for rep_file in sorted(model_dir.glob("rep*.json")):
-            data = json.loads(rep_file.read_text())
+            data = json.loads(rep_file.read_text(encoding="utf-8"))
             for rec in data["records"]:
                 if rec.get("error"):
                     # count errored task-runs as score 0 but keep them in the
--- a/evals/session_search_schema/runner.py
+++ b/evals/session_search_schema/runner.py
@@ -50,7 +50,7 @@ def _load_api_key() -> str:
         return key
     env_path = Path.home() / ".hermes" / ".env"
     if env_path.exists():
-        for line in env_path.read_text().splitlines():
+        for line in env_path.read_text(encoding="utf-8").splitlines():
             if line.startswith("OPENROUTER_API_KEY="):
                 return line.split("=", 1)[1].strip().strip('"').strip("'")
     raise SystemExit("OPENROUTER_API_KEY not found (env or ~/.hermes/.env)")
@@ -65,7 +65,7 @@ def extract_arm(ref: str, workdir: Path, name: str) -> Path:
     if out.returncode != 0:
         raise SystemExit(f"git show {ref}: {out.stderr.strip()}")
     path = workdir / f"ss_arm_{name}.py"
-    path.write_text(out.stdout)
+    path.write_text(out.stdout, encoding="utf-8")
     return path
 
 
@@ -237,7 +237,7 @@ def main():
         outpath = outdir / (re.sub(r"[^\w.-]", "_", args.model) + ".jsonl")
         done = set()
         if outpath.exists():
-            for line in outpath.read_text().splitlines():
+            for line in outpath.read_text(encoding="utf-8").splitlines():
                 try:
                     r = json.loads(line)
                     done.add((r["task"], r["arm"], r["rep"]))
--- a/scripts/check-windows-footguns.py
+++ b/scripts/check-windows-footguns.py
@@ -113,6 +113,17 @@ EXCLUDED_FILES = {
     "CONTRIBUTING.md",
 }
 
+# Directories --all walks past. Unlike EXCLUDED_DIRS these are still scanned
+# when named explicitly or reached through --diff / staged mode, so a footgun
+# introduced in a test is still caught at review time. `tests/` is skipped by
+# the full sweep only because it is overwhelmingly the larger half of the repo
+# and currently carries ~2.3k matches (nearly all bare read_text()); letting
+# them into the blocking CI sweep would bury the handful of production hits
+# this check exists to surface. Cleaning them up is its own campaign.
+FULL_SCAN_SKIP_DIRS = {
+    "tests",
+}
+
 
 @dataclass
 class Footgun:
@@ -456,6 +467,24 @@ def should_scan_file(path: Path) -> bool:
     return False
 
 
+def full_scan_roots() -> list[Path]:
+    """Top-level entries --all sweeps: the whole repo minus the skip lists.
+
+    Returns repo-root children rather than REPO_ROOT itself so the top-level
+    skip decision is explicit and testable; ``iter_files`` still applies
+    EXCLUDED_DIRS / EXCLUDED_SUFFIXES / EXCLUDED_FILES underneath each one.
+    Top-level *files* (cli.py, conftest.py, ...) are included — they were
+    invisible to the old package list.
+    """
+    return sorted(
+        child
+        for child in REPO_ROOT.iterdir()
+        if not child.name.startswith(".")
+        and child.name not in EXCLUDED_DIRS
+        and child.name not in FULL_SCAN_SKIP_DIRS
+    )
+
+
 def iter_files(paths: Iterable[Path]) -> Iterable[Path]:
     for p in paths:
         if p.is_file():
@@ -709,7 +738,10 @@ def parse_args(argv: list[str]) -> argparse.Namespace:
     p.add_argument(
         "--all",
         action="store_true",
-        help="Scan the full repository (hermes_cli/, gateway/, tools/, cron/, etc.).",
+        help=(
+            "Scan the full repository (everything except "
+            f"{'/, '.join(sorted(FULL_SCAN_SKIP_DIRS))}/ and the always-excluded dirs)."
+        ),
     )
     p.add_argument(
         "--diff",
@@ -749,18 +781,14 @@ def main(argv: list[str]) -> int:
         return 0
 
     if args.all:
-        # Scan main Python packages + scripts
-        roots = [
-            REPO_ROOT / "hermes_cli",
-            REPO_ROOT / "gateway",
-            REPO_ROOT / "tools",
-            REPO_ROOT / "cron",
-            REPO_ROOT / "agent",
-            REPO_ROOT / "plugins",
-            REPO_ROOT / "scripts",
-            REPO_ROOT / "acp_adapter",
-        ]
-        roots = [r for r in roots if r.exists()]
+        # Walk the repository rather than a hand-maintained package list. The
+        # old list named eight directories and silently skipped everything
+        # else — tui_gateway/, evals/, providers/, skills/ and the top-level
+        # modules like cli.py — so the blocking CI sweep reported "no footguns
+        # found" for code it had never opened. Anything new at the repo root
+        # is now covered the day it lands instead of the day someone
+        # remembers to extend this list.
+        roots = full_scan_roots()
     elif args.diff:
         roots = get_diff_files(args.diff)
     elif args.paths:
--- a/tests/scripts/test_windows_footguns_full_repo_scan.py
+++ b/tests/scripts/test_windows_footguns_full_repo_scan.py
@@ -16,12 +16,30 @@ the stdin guard already closes its equivalent one.
 
 from __future__ import annotations
 
+import importlib.util
 import subprocess
 import sys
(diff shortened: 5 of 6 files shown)


alt-glitch:

Duplicate of #85633: both replace the --all hardcoded roots with full-repository coverage for the Windows-footgun scanner.

BkashJEE:
Thanks — #85633 does predate this and I should have found it. I searched the symptom terms (`os.kill liveness`, `conhost`, `windowsHide`) and never the fix terms (`footgun`, `--all`), which is on me.

They are not quite the same change though, and the difference decides whether #97019 gets fixed, so flagging it for whoever triages rather than assuming:

**#85633 extends the hardcoded list — but not with `tui_gateway/`.** It adds `evals`, `website`, `providers`, `skills` and a top-level `*.py` glob. `tui_gateway` is not among them (`gh pr diff 85633 | grep -c tui_gateway` → 0). That is the directory holding the live `os.kill(pid, 0)` from #97019 and a bare `signal.SIGKILL`, so after #85633 the sweep still reports clean over the P2 bug. It also fixes only `evals/readtool/report.py`, using `encoding="utf-8-sig"` rather than `"utf-8"` — BOM-stripping rather than the plain UTF-8 the rule's own `Fix:` line suggests.

**This PR replaces the list with a repo walk.** No list to maintain, so `tui_gateway/` and anything added at the root later are covered by construction, and it fixes all seven findings that surface — including the two in `host_supervisor.py` — so the blocking job stays green. `tests/` is skipped by the sweep alone (it carries ~2.3k matches, nearly all bare `read_text()`); explicit paths, `--diff` and staged mode still scan it.

Either way #85633 got there first and has been sitting since 2026-08-13 with no CI run recorded. If you would rather land that one, the useful merge is probably: take its walk plus `tui_gateway` and the seven fixes from here, or just add `REPO_ROOT / "tui_gateway"` to its list and fix the two `host_supervisor.py` sites — otherwise #97019 stays open behind a green checker. Happy to rebase this onto #85633, cut it down to only what that PR misses, or close it entirely — whichever is least work for you.

BkashJEE:
Closing this. A proper search — the one I should have run before opening it — turns up prior work covering every part of this PR:

- `--all` coverage: #85633 (2026-08-13)
- `host_supervisor._pid_alive` / `os.kill(pid, 0)`: #68272 (08-10), #84323 (08-16), #90540 (08-21), #97022 (08-28), #97187 (08-28)

#97022 in particular is this PR's `host_supervisor` change done better: same routing through `gateway.status._pid_exists`, it fixes the `signal.SIGKILL` escalation site as well, and it adds a dedicated regression module that pins the probe against `gateway.status._pid_exists` rather than `os.kill`, following `TestOwnerAlivePidProbe` in `tests/gateway/test_delivery_ledger.py`. Nothing here improves on it.

The one observation I have not seen made elsewhere, in case it is useful to whoever lands #85633: that PR extends the `--all` root list with `evals`, `website`, `providers`, `skills` and a top-level `*.py` glob, but not `tui_gateway`. So after it merges the blocking sweep still walks past `tui_gateway/host_supervisor.py`. Adding `REPO_ROOT / "tui_gateway"` to its list, or replacing the list with a repo walk, would close that. Otherwise the checker stays green over whichever `os.kill(pid, 0)` PR has not landed yet.

Measured while working on this, if any of it saves someone time:

- `--all` scans 1021 files; a repo walk minus `tests/` scans 1168. The 147-file delta contains 7 footguns: 2 in `tui_gateway/host_supervisor.py`, 3 in `evals/session_search_schema/runner.py`, 1 in `evals/readtool/report.py`, 1 in `apps/desktop/scripts/perf/gateway_attach_bench.py`.
- `tests/` itself holds ~2.3k matches. ~2,152 are bare `read_text()`/`write_text()`; the other ~55 are behavioural — 19 `os.killpg`, 12 `os.getuid`/`geteuid`, 10 `signal.SIGKILL`, 8 `os.kill(pid, 0)`, 6 `os.setsid`.

Apologies for the duplicate. My search used symptom vocabulary rather than fix vocabulary, which is exactly how you miss five prior PRs on the same function.

PR #88499, opened 2026-08-17
fix(desktop): align sidebar density row heights

enwaiax:
## Summary

- align the virtual session-list estimates with explicit rendered minimum heights for every density
- raise comfortable rows to 48px and detailed rows to 64px
- keep compact rows at their existing one-line 28px height
- centralize the virtualizer height contract in `SESSION_ROW_HEIGHT_PX`

Closes #88473

## Root cause

The virtualized sidebar positions rows absolutely from `sessionRowEstimate()`, while the rendered row used different Tailwind minimum heights. Comfortable rendered at 44px but was placed 45px apart; detailed rendered at 62px but was placed 63px apart. Browser text metrics and padding could make the content taller than either value, allowing adjacent rows and date dividers to overlap before or between measurement passes.

The renderer and virtualizer now use matching 48px/64px boundaries with enough room for their two- and three-line content. Compact remains a one-line 28px row.

## Verification

- `npm exec --workspace=apps/desktop vitest -- run --project ui src/app/chat/sidebar/session-row-details.test.ts src/app/chat/sidebar/virtual-session-list.test.tsx` — 10 passed
- `npm run typecheck --workspace=apps/desktop`
- `npm run lint --workspace=apps/desktop -- --quiet`
- `git diff --check`

## Scope

This does not change virtualization, scrolling, density selection, or session data. It only makes row placement and rendered geometry agree.

Issue #88473, opened 2026-08-17
[Bug]: Desktop sidebar session rows overlap (density row-height estimate too small)


igorl-commits:
## Bug Description

Session rows in the desktop sidebar overlap each other. Each row paints over the row below it. Preview text covers the title and metadata of the next row. Date dividers (YESTERDAY, LAST WEEK) are covered by adjacent rows.

The sidebar uses a virtualized list. Each row is positioned with `translateY(start)`. The list sets the position from an estimated row height. The estimate is too small for the actual rendered row. So the list places rows too close together and they overlap.

## Steps to Reproduce

1. Open Hermes Desktop.
2. Open the sidebar Sessions view.
3. Have 25+ sessions so the list virtualizes.
4. Look at the session rows.

## Expected Behavior

Each session row has enough vertical space. Row content does not overlap the next row.

## Actual Behavior

Session rows overlap. Previews cover titles and metadata. Date dividers are covered.

## Environment

- Operating System: Windows 11
- Hermes Version: Desktop 0.17.0, backend 0.20.2
- Head commit: 4323c67dcc

## Root Cause Analysis

The bug came from commit `a33ac1e95e` (feat: add session list density modes, merged 2026-08-15).

`apps/desktop/src/app/chat/sidebar/session-row-details.ts`

```ts
export const sessionRowEstimate = (density) =>
  ({ compact: 28, comfortable: 45, detailed: 63 })[density]
```

`apps/desktop/src/app/chat/sidebar/virtual-session-list.tsx`

The virtualizer uses `estimateSize` to place each row. Rows are absolutely positioned with `translateY(start)`.

In `compact` mode (the default) the row renders a title, a preview, a Home label, and a metadata line. This is about 60px tall. The estimate is 28px. The list places each row 28px apart. The actual row is 60px tall. So each row overlaps the next.

In `comfortable` mode the row adds `min-h-[2.75rem]` (44px) and `detailed` adds `min-h-[3.875rem]` (62px). The estimates are 45px and 63px. These still do not match the rendered height exactly, so overlap persists in all modes.

Changing density calls `virtualizer.measure()` (line 107 of `virtual-session-list.tsx`). This reduces the overlap but does not remove it.

## Proposed Fix

Make the row-height estimate match the rendered height in every density mode.

Option A: Add a minimum height to the `compact` row, as the other modes have, and raise the `compact` estimate to that height.

[...]

PR #97353, opened 2026-08-28
fix(desktop): prevent Inbox session row overlap (#88473)

jorgemdnt:
## What does this PR do?

Prevents Inbox-style cards in the virtualized flat Sessions list from painting over adjacent sessions and date dividers.

The compact-to-Inbox transition retained compact row measurements because the invalidation effect only watched density. If a resize observation was stale or missed, TanStack Virtual then fell back to a 74px card estimate for cards that render at about 92px.

The fix stays inside the existing virtualization contract:

- remeasure when `card` mode changes;
- use 92px as the full-card fallback estimate while normal `measureElement` calls still tighten shorter rows.

No row CSS, density behavior, scroll chaining, row identity, or Project-grouping path changes.

## Related issue

Addresses the **Inbox style + Updated/Status/Profile** case documented in #88473.

This is separate from #88499: that PR aligns comfortable/detailed non-card density heights; it does not change card-mode invalidation or the card estimate.

## Changes made

- Update `VirtualSessionList` to invalidate cached measurements when Inbox card mode toggles.
- Raise the card fallback from 74px to the measured four-line card height of 92px.
- Add unit coverage for both compact → Inbox and Inbox → compact invalidation.
- Add a real Electron regression with 28 durable sessions, an interleaved date divider, a missed resize delivery, off-screen scrolling, and both mode transitions.

### Red / control / green

| Run | Result |
|---|---|
| Upstream target code | Electron test fails; rows overlap by up to **65.74px** |
| Causal control | Resuming resize delivery and forcing a card resize clears overlap to **0px** |
| `2e478893c` | Same Electron test passes with resize delivery still suppressed; **0px** overlap |

### Checks on exact head `2e478893c`

- `npx playwright test e2e/inbox-card-measurement.spec.ts --reporter=line` — 1 passed
- `npx vitest run --project ui src/app/chat/sidebar` — 24 files, 221 tests passed
- `npm run typecheck` — all three desktop TypeScript configs passed
- `npx eslint src/ electron/ e2e/inbox-card-measurement.spec.ts --quiet` — clean
- `npm run build` — clean packaged build; install stamp pinned to the exact SHA
- `git diff --check` — clean

Python tests were not run because this PR changes only the desktop TypeScript/Electron surface.

Tested on macOS 15.6 (Apple Silicon). The E2E uses Node's built-in SQLite API rather than a platform-specific CLI and will also run in the Linux desktop CI job.

## Screenshots

Both captures use the same 28-session Electron fixture and 520×1436 sidebar viewport.

| Before — upstream target code, 65.74px overlap | After — `2e478893c`, 0px overlap |
|---|---|
| Before: Inbox session cards and date divider overlapping | After: Inbox session cards and date divider separated |

### DIFF
--- a/apps/desktop/e2e/inbox-card-measurement.spec.ts
+++ b/apps/desktop/e2e/inbox-card-measurement.spec.ts
@@ -1,0 +1,261 @@
+/**
+ * Electron regression for stale virtual-row measurements when the session
+ * sidebar switches from compact rows to Inbox cards.
+ */
+
+import * as path from 'node:path'
+import { DatabaseSync } from 'node:sqlite'
+
+import type { Page } from '@playwright/test'
+
+import {
+  buildAppEnv,
+  createSandbox,
+  launchDesktop,
+  type MockBackendFixture,
+  waitForAppReady,
+  writeEnvFile,
+  writeMockProviderConfig
+} from './fixtures'
+import { startMockServer } from './mock-server'
+import { RealSessionBuilder } from './real-session-builder'
+import { expect, test } from './test'
+
+const SESSION_COUNT = 28
+const SESSION_PREFIX = 'E2E Inbox measurement row'
+const OVERLAP_TOLERANCE_PX = 0.5
+const RESIZE_GATE = '__hermesE2EDropResizeObservations'
+
+interface SessionRowRect {
+  bottom: number
+  height: number
+  index: number
+  session: boolean
+  text: string
+  top: number
+}
+
+interface LayoutSample {
+  maxOverlap: number
+  overlaps: Array<{ current: number; next: number; pixels: number }>
+  rows: SessionRowRect[]
+}
+
+async function setupSeededSidebar(): Promise<MockBackendFixture> {
+  const mock = await startMockServer()
+  const sandbox = createSandbox('inbox-card-measurement')
+
+  try {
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Fixes stale `react-virtuoso` row measurements when sidebar switches from compact rows to Inbox cards (#88473) and adds Electron regression coverage.

**What it does**
- `apps/desktop/src/app/chat/sidebar/virtual-session-list.tsx` fix: invalidates virtual-row height cache on Inbox vs compact mode switch so card height is remeasured (prevents overlap).
- `virtual-session-list.test.tsx` asserts remeasure.
- `apps/desktop/e2e/inbox-card-measurement.spec.ts` new 261-line Playwright/Electron spec: seeds 28 sessions via `RealSessionBuilder` + `DatabaseSync`, toggles to Inbox, samples `getBoundingClientRect` via `LayoutSample`, asserts `maxOverlap <= 0.5px` and no resize observation leak (`__hermesE2EDropResizeObservations` gate).

**Non-blocking notes**
- E2e seeds 28 sessions — heavy but bounded; ensure CI timeout covers `startMockServer` + builder close. Overlap tolerance 0.5px accounts for subpixel rounding — appropriate.
- Gate name `__hermesE2EDropResizeObservations` is test-only — verify stripped in production build.

No blocking issues.

PR #43669, opened 2026-06-10
Suppress "Credit access paused" notice on free models

rob-maron:
It currently causes confusion when a free or paid-with-no-credits user goes to use a free model

PR #97354, opened 2026-08-28
fix(nous): honour the org model policy in the model pickers

entropidelic:
## What does this PR do?

A Nous team admin can restrict which models and which serving providers their org may use, and the inference gateway now applies that policy to `GET /v1/models` (api#164), omitting blocked rows with no marker field. Hermes ignored it: every surface built its Nous list from the docs-hosted curated manifest unioned with the Portal's `recommended-models` endpoint, neither of which is authenticated. Users were offered models their org blocks, then got `403 model_blocked_by_org_policy` on the first request.

Four surfaces now narrow their list to what the org can actually reach, and a governed org is told its choice is restricted rather than silently seeing a shorter list.

The approach reuses the authenticated `GET /v1/models` read that every surface **already** makes for pricing, rather than adding a catalog call. Net effect is one request *fewer* per picker open, and ~2 ms of added work. Filtering is gated on the token's `policy_present` claim, so an org without a policy takes no catalog read and its list is byte-identical to today — a no-op for everyone the policy doesn't apply to.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `hermes_cli/models.py` — `nous_policy_allowed_ids()` reuses the authenticated pricing response's keys as the reachable set; `restrict_to_nous_policy()` applies it while preserving curated order. Both decline to narrow when the evidence can't support it: no policy, an anonymous (therefore unfiltered) read, or an empty read. Also fixes `fetch_models_with_pricing`, whose cache keyed on base URL alone with the check sitting *above* the point where the `Authorization` header is built. It now keys on a blake2b fingerprint of the credential, so one org's policy-filtered catalog can't answer another's read after a profile switch or re-login, and Nous entries carry a 5-minute TTL so a policy change lands without a process restart. Both are opt-in for Nous: other providers keep no-expiry, URL-keyed caching.
- `hermes_cli/nous_account.py` — `nous_policy_present()` reads the `policy_present` claim off the access token, no request. Tri-state: an absent claim is unknown, not "no policy". `nous_policy_notice()` returns the one-line notice.
- `hermes_cli/auth.py` (`_login_nous`), `hermes_cli/model_setup_flows.py` (`_model_flow_nous`), `hermes_cli/model_switch.py` (`list_authenticated_providers`), `hermes_cli/web_server.py` (`/api/model/recommended-default`) — narrow the list at all four surfaces; the two interactive flows also print the notice. The `/model` picker's filter sits outside the `try` wrapping the Portal union, so a Portal outage still yields a filtered curated list.
- `agent/auxiliary_client.py` — `_fast_model_from_catalog` used the catalog's keys as a source of ids off an *anonymous* read, because the api-key resolver raises for OAuth providers. It now resolves Nous credentials, and the policy check moved to cover the whole resolution ladder: with no fast-family match in the allowed catalog it fell through to a public recommendation, which could hand titling a blocked model. A blocked pick now yields no aux model, so the caller keeps the main one.
- `hermes_cli/model_switch.py` — Nous dropped from the background prefetch. Its picker branch builds from the curated list and it can't reach the api_key-only pathway, so the prefetched entry was written and never read: a live authenticated `/v1/models` round trip per picker open for nothing.
- `agent/credits_tracker.py` — reads the pricing cache via `peek_cached_pricing()` instead of duplicating the cache-key format.

Two behaviours worth flagging for review, both found while testing against a preview:

- **An allowlist can name only models the curated manifest lacks** (e.g. `amazon/nova-2-lite-v1`). Intersection alone turns that into an empty picker — worse than no filter, since the models the org may use are exactly the ones dropped. The selectable list opts into falling back to the reachable set in that case; no other list does, because an already-empty list (a paid tier's gated models) means "nothing to gate", not "nothing survived". The narrowing also runs *before* the free/paid split, so a rescued id still has to pass the tier predicate rather than going around it.
- **`policy_present` is stamped at token mint time**, so a newly-applied policy isn't reflected client-side until the next token refresh, bounded by the catalog TTL above. The gateway still refuses the request with a 403, so it's a display lag, not a bypass.

## How to Test

The NAS and gateway changes are in production, so this works against a normal Nous login — no env overrides needed. You need an org whose policy you can edit.

1. `uv run hermes model` → pick Nous → complete the device-code login if needed.
2. In the Nous Portal, set a **model allowlist** naming one model (e.g. `amazon/nova-2-lite-v1`), then re-run `uv run hermes model`. Expect that single model plus `Your organization restricts which models are available — models outside its policy are not listed.`
3. Swap it for a **jurisdiction/provider policy** (e.g. CHINA). Expect the curated list narrowed to models those providers serve, in curated order, with the same notice.
4. Clear the policy. Expect the full curated list, no notice, and no extra catalog request.

`policy_present` is stamped at token mint time, so after changing a policy you may need to re-login (or wait for the access token to refresh) before the client reflects it. The gateway enforces it regardless.

To test against staging instead, set the overrides in `~/.hermes/.env` and re-login — they bypass the host allowlist, which rejects non-production hosts persisted in `auth.json`:

```
HERMES_PORTAL_BASE_URL=https://<nas-staging>
NOUS_INFERENCE_BASE_URL=https://<gateway>/v1
```

Isolating client from server, if a step disagrees:

```bash
uv run python -c "
from hermes_cli.env_loader import load_hermes_dotenv; load_hermes_dotenv(load_external_secrets=False)
from hermes_cli.nous_account import nous_policy_present
from hermes_cli.models import _resolve_nous_pricing_credentials, fetch_models_with_pricing
k, b = _resolve_nous_pricing_credentials()
print('policy_present:', nous_policy_present(), '| authenticated:', bool(k))
a = fetch_models_with_pricing(api_key=k, base_url=b, include_sale_original=True)
n = fetch_models_with_pricing(api_key='', base_url=b, include_sale_original=True)
print('authed rows:', len(a), '| anon rows:', len(n), '| blocked:', len(set(n) - set(a)))
"
```

`blocked: 0` with `authenticated: True` means the gateway isn't filtering (server-side); `policy_present: False`/`None` means the token carries no policy claim, so the client deliberately leaves the list alone.

**Observed against a NAS preview:** jurisdiction=CHINA → gateway returned 51 of 343 rows, picker showed 8 curated models (all served by those providers), `unavailable_models` empty. Allowlist of `amazon/nova-2-lite-v1` alone → picker showed that one model. Ungoverned org → zero catalog reads, list unchanged.

### Code

- [x] I've tested on my platform: macOS 15 (Darwin 24.6.0), Python 3.11

56 new tests across 3 files; 358 green across the 8 existing suites that touch these paths. Every filter confirmed by negative control — removing it fails its test. `ruff check` and `scripts/check-windows-footguns.py --all` clean. Footprint 11 files (+1010/−27).


alt-glitch:

@rob-maron Tagging you on this credits-notice item as the soft maintainer (follow-up to #43669).

agent/auxiliary_client.py
```diff
@@ -884,6 +884,17 @@ def _fast_model_from_catalog(provider_id: str) -> str:
...
             logger.debug("No credentials for %s catalog", provider_id, exc_info=True)
 
+        if not api_key and provider_id.strip().lower() == "nous":
+            # Nous is OAuth, so the resolver above raises for it. An anonymous
+            # read returns the full catalog, and a model picked from it is
+            # refused at request time by the org's policy.
+            try:
+                from hermes_cli.models import _resolve_nous_pricing_credentials
+
+                api_key, base_url = _resolve_nous_pricing_credentials()
```
pefontana, line 895: Pass cache_ttl_seconds=_NOUS_CATALOG_TTL_SECONDS and include_sale_original=True to the fetch below, otherwise when titling runs first it seeds this shared cache entry with no expiry and no sale pricing.

hermes_cli/models.py
```diff
@@ -2242,19 +2242,65 @@ def _cached_catalog(cache_key: str) -> Optional[dict[str, dict[str, Any]]]:
...
+
+    Accepts a ``/v1``-suffixed URL as well as the pre-``/v1`` root the fetchers
+    key on, and prefers an authenticated catalog. Scans rather than rebuilding a
+    key, because callers hold a base URL but no credential.
+    """
+    root = (base_url or "").rstrip("/")
+    if root.endswith("/v1"):
+        root = root[:-3].rstrip("/")
+    authed_prefix = root + _PRICING_AUTH_KEY_PREFIX
+    for key, cached in _pricing_cache.items():
```
pefontana: Skip entries whose _pricing_cache_retry_after deadline has passed and prefer the newest match, otherwise after a credential rotation this keeps returning the old catalog.

hermes_cli/nous_account.py
```diff
@@ -396,6 +396,46 @@ def get_nous_portal_account_info(
...
+    """A one-line notice for an org that restricts model choice, else ``""``.
+
+    A blocked model is omitted rather than marked, which reads as "Hermes does
+    not support this". This says which it is without enumerating the blocked
+    set, which under an allowlist is most of the catalog.
+    """
+    if nous_policy_present() is not True:
+        return ""
+    return (
+        "Your organization restricts which models are available — "
```
pefontana, line 438: The gateway catalog read fails open, so show this only when the filter actually removed something, otherwise it can label a full list as filtered.

pefontana (approved):

PR #88487, opened 2026-08-17
fix(gateway): scope projects.* RPCs to the active profile

TomSpoct:
## Problem

In **app-global remote mode**, one dashboard backend serves every Hermes profile over a single WebSocket. The `projects.*` JSON-RPC handlers ignored `params['profile']` entirely and always opened the **launch (default) profile's** `projects.db` / `state.db`.

The desktop then grouped every profile's sessions using the default profile's Projects. Symptom (reported by user): sessions in the **Ray** profile appeared under default-profile Projects like **RSS Dashboard** and **Emma General**, even though Ray's own `projects.db` was empty.

### Evidence

- `hermes --profile ray project list` → "No projects yet"
- Ray `state.db` → 54 sessions with `cwd = NULL` (should land in the Home bucket), 1 with `cwd = /home/tom/workspaces/emma`
- Default `projects.db` → `rss-dashboard` + `emma-general`
- Desktop Project RPCs (`projects.list`, `projects.tree`, `projects.project_sessions`, `projects.record_repos`) sent **no** `profile` param

## Fix

**Backend** (`tui_gateway/`):

- `_projects_method` (CRUD + `projects.list`) now resolves the owning profile's `projects.db` via `_profile_home(profile)` from `params['profile']`.
- `projects.tree`, `projects.project_sessions`, `projects.record_repos`, `projects.discover_repos` now read the profile's `state.db` through the existing `_profile_db(params)` context manager and open the profile's `projects.db` through the scoped path.
- Launch/own profile (or omitted `profile`) keeps the existing unscoped path — same DB, no override.

**Desktop** (`apps/desktop/src/store/projects.ts`):

- Every `projects.*` request now stamps the active gateway profile (`projectRpc` / `projectRpcOn` helpers), so the shared backend can route it. Mirrors the existing pet-RPC helper pattern. The launch-profile stamp is a no-op on the backend.

## Tests

- Backend: `test_projects_rpc_is_profile_scoped` — seeds a named profile's `projects.db`, asserts `projects.list` with `profile` reads only that profile, and that scoped writes never land in the launch profile.
- Desktop: `projects RPC profile stamping` suite — asserts `projects.list`, `projects.tree`, `projects.project_sessions` carry the active profile.

Verification: `pytest tests/tui_gateway/test_projects_rpc.py` (24 passed), `pytest tests/test_tui_gateway_server.py` (585 passed), `vitest run src/store/projects.test.ts` (31 passed), `tsc -p tsconfig.json --noEmit` clean.

PR #81478, opened 2026-08-08
fix(tui_gateway): profile-scope the projects.* RPC surface

tsemana:
### Problem

In the desktop's app-global remote mode, one backend serves every local
profile. `session.*` RPCs handle this: the desktop sends a `profile` param
and the handlers rebind `HERMES_HOME` (ContextVar) plus open that profile's
`state.db` for the duration of the call.

The `projects.*` surface never got the same treatment, on either end:

- Backend: `projects.tree` / `projects.project_sessions` /
  `projects.discover_repos` / `projects.record_repos` read sessions from the
  module-global launch-profile `_get_db()` handle, and the `_projects_method`
  CRUD family resolves `projects.db` via `connect_closing()` without any
  profile binding — so both always hit the **launch** profile's databases.
- Renderer: the projects store sends no `profile` field on any projects
  request (`apps/desktop/src/store/projects.ts`), even though the file's own
  comment says "Projects are per-profile, so they intentionally follow the
  active gateway just like the session list does."

**Repro**: create a second profile, add a project to it, switch the desktop
to that profile. The session lanes rescope correctly; the projects sidebar
still shows the launch profile's tree, and the second profile's projects are
unreachable from the UI.

### Fix

Mirrors the existing `session.*` pattern exactly:

- The four tree/discovery handlers in `tui_gateway/methods_config.py` take
  `@_profile_scoped` and read sessions via `with _profile_db(params) as db:`
  instead of `_get_db()`.
- `_projects_method` in `tui_gateway/server.py` applies `@_profile_scoped`
  once for the whole CRUD family — `projects_db_path()` already resolves
  through `get_hermes_home()`, so the ContextVar rebind is sufficient there.
- The renderer's projects store merges the active gateway profile into every
  projects request (`activeProfileParam()` in both request helpers). The
  param is omitted for `default`, so single-profile installs send
  byte-identical requests and the backend path for them is unchanged. Callers
  already race-guard profile switches (generation counter + active-gateway
  check), so tagging with the active profile at send time is safe.

### Compatibility

- Single-profile installs: no behavior change (no `profile` sent; backend
  treats a missing param as the launch profile, as before).
- Old renderer + new backend, or new renderer + old backend: degrade to
  today's behavior (launch-profile tree) rather than erroring.

### Testing

- Manually verified on a macOS install with two profiles served by one
  remote-mode backend: the projects sidebar now switches with the profile
  rail (each profile shows its own `projects.db` tree + session grouping);
  the launch profile's view is unchanged.
[...]

PR #97355, opened 2026-08-28
fix(desktop): resolve session projects from owning profile

evan-bradford:
## Summary

Resolve Desktop session project metadata from the session's owning profile rather than the backend process's launch profile.

A multiplexed Desktop backend can serve sessions from several profile homes, each with its own `projects.db`. Outside a profile-scoped RPC, ambient project lookup still points at the launch profile. That allowed a secondary-profile session to inherit a same-path project from the wrong profile.

## Root cause

`_project_info_for_cwd()` always opened the ambient `projects.db`, even when the session already carried its owning `profile_home`.

## Fix

- Let `_project_info_for_cwd()` accept an optional profile home.
- Open `<profile_home>/projects.db` when ownership is known.
- Thread session ownership through live, lazy-resume, fallback, workspace-update, immediate-create, cwd-set, workspace-move, and status metadata paths.
- Preserve existing ambient behavior for profile-less/single-profile callers.

## Regression coverage

The tests create conflicting default-profile and secondary-profile project registries for the same working directory, then verify the owning project through:

- `session.info`
- immediate `session.create` metadata
- lazy `session.cwd.set`
- live `session.workspace.move` event metadata
- `session.status` output

## Verification

- `scripts/run_tests.sh tests/tui_gateway/test_projects_rpc.py tests/tui_gateway/test_session_profile_db.py -q` — **44 passed**
- Existing fallback project-info compatibility tests — **2 passed**
- Ruff, diff check, and added-line security scan — passed
- `git diff --check origin/main...HEAD` — passed
- Initial independent review caught four modular handler call sites omitted from the first patch; all four were reproduced RED through public RPCs and are GREEN in follow-up `2018a760e`
- A filesystem-aware re-review confirmed all eight live call sites, then caught two stale one-argument test stubs; compatibility is restored in `92c90c5bc`
- Independent filesystem-aware final-head review of `92c90c5bc` — **PASS, no blockers**

## Scope

This changes display/project metadata lookup only. It does not mutate conversation history, toolsets, prompts, or profile configuration.

### DIFF
--- a/tests/test_tui_gateway_server.py
+++ b/tests/test_tui_gateway_server.py
@@ -19873,7 +19873,9 @@ def test_fallback_session_info_reports_session_cwd_not_launch_dir(monkeypatch):
     """
     monkeypatch.setattr(server, "_default_session_cwd", lambda: "/gateway/launch/dir")
     monkeypatch.setattr(server, "_git_branch_for_cwd", lambda cwd: "bb/feature")
-    monkeypatch.setattr(server, "_project_info_for_cwd", lambda cwd: None)
+    monkeypatch.setattr(
+        server, "_project_info_for_cwd", lambda cwd, profile_home=None: None
+    )
     monkeypatch.setattr(server, "_resolve_model", lambda: "test-model")
 
     info = server._fallback_session_info({"cwd": "/projects/session-own-repo"})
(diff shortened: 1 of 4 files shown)


alt-glitch:

Related: #88487 and #81478 address the broader projects-RPC profile-scoping cluster. This PR fixes the distinct session.info project-metadata lookup path.

evan-bradford:
## Live deployment verification — 2026-08-28

I deployed the reviewed changes as part of combined commit `96d0deb87d8726ed784e83595177bf2a42c70b1b`, restarted the real Spark dashboard backend after that commit, and ran a read-only JSON-RPC verification against the original failure case.

For the actual session titled `Complete highest-leverage internal GTM work`:

- exact-title lookup in profile `sfb`: resolved session `20260827_153918_4d4408`
- exact-title lookup in profile `default`: **0 rows**
- appearances in `projects.project_sessions {profile: "sfb"}`: **exactly 1**
- returned owner stamp: `profile: "sfb"`
- returned CWD: `/home/evanbradford/Projects/Songs-for-Brands/songs-for-brands-strategy/.worktrees/gtm-strategy-08212026`

The live backend reports `overall: ok`, `gateway_state: running`, with all dashboard/gateway/storage/platform components healthy. This confirms the owning-profile project registry is used on the real multiplexed backend, not only in unit fixtures.

PR #95473, opened 2026-08-26
fix(desktop): look up v1 SSH terminals with the bootstrap pool key

Adolanium:
## What does this PR do?

#95081 keyed SSH tunnels per connection. v2 registry windows look up `sshConnections` with `backendScopeKey(connectionId, profile)`. That is correct for registry-scoped windows.

v1 settings/profile SSH still bootstraps with `sshScopeKey(profile or null)`. `resolveDesktopRemoteRoute` also stamps a registry `connectionId` when the host matches a v2 row. `activeSshTerminalTarget` then used that id, so the lookup was `conn:<id>::<profile>` while the live tunnel sat under `''` or the profile name. The embedded terminal stayed pending.

v1 pool lookup now matches bootstrap and ignores the identity tag. The v2 registry-scoped branch at the top of `activeSshTerminalTarget` is unchanged.

## Related Issue

Fixes #95472

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `apps/desktop/electron/desktop-remote-route.ts`: `v1SshTerminalPoolKey` returns `''` for settings SSH and the profile name for profile SSH.
- `apps/desktop/electron/main.ts`: `activeSshTerminalTarget` uses that helper on the v1 path.
- `apps/desktop/electron/desktop-remote-route.test.ts`: settings SSH with a matching registry id still pools under `''`. Profile SSH pools under the profile name, not `conn:id::profile`.

## How to Test

1. From `apps/desktop`: `npx vitest run electron/desktop-remote-route.test.ts`
2. v1 settings SSH whose host matches a registry row must have `connectionId` set and `v1SshTerminalPoolKey` still `''`.
3. v1 profile SSH must pool under the profile name, not `backendScopeKey(connectionId, profile)`.

I could not run the desktop vitest project in this checkout (no `node_modules`). The new tests are next to the existing route tests.

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
[...]

PR #97356, opened 2026-08-28
fix(desktop): resolve v2 registry SSH terminals from the bootstrap pool key

### LINKED ISSUE #97345: [Bug]: Desktop terminal pane never starts on a v2 registry SSH connection — `activeSshTerminalTarget` reads a `backendScopeKey` the bootstrap writer never writes (regression from #95081)
### Summary

After #95081 (`71d804d00`), the desktop terminal pane on a **v2 registry SSH connection** fails instantly on every tab and every relaunch:

```
Terminal failed to start: Error invoking remote method 'hermes:terminal:start':
Error: Remote connection is not ready yet. Try again in a moment.
```

Chat, sessions, profiles and tools all work normally over the same SSH tunnel at the same moment. Only the terminal pane fails.

The cause is a **read/write key mismatch on the `sshConnections` pool**. `activeSshTerminalTarget()` looks the tunnel up under a composite `backendScopeKey(connectionId, profile)`, but the one and only writer stores it under `sshScopeKey(profile)`. The lookup can never hit, so the pane is permanently `'pending'`.

This is the same *class* of defect as #95472, but a **different branch of the same function**, and PR #95473 explicitly does not touch it. See "Relationship to other issues" below.

### Environment

| | |
|---|---|
| Hermes Desktop | v0.20.6 (2026.8.27), Windows 11 |
| Remote backend | Hermes Agent v0.20.6, Linux x86_64, over SSH |
| Client checkout | `1acd5bb02bbae0b32416a3ac45d4f00373453e33` (contains `71d804d00`; `git merge-base --is-ancestor` exit 0) |
| Connection | `connections.json` v2, `primary: "<conn-id>"`, `launchMode: "primary"`, `kind: "ssh"` |
| Verified against | `origin/main` — the code below is byte-identical on `main` today |

### Root cause

**There is exactly one writer to the `sshConnections` pool on `main`.** `apps/desktop/electron/main.ts:10269`, inside `bootstrapSshConnectionInner()`, whose scope comes from line 10158:

```ts
// main.ts:10158
const scope = sshScopeKey(profile)      // profile-only key -> "" or "default"
...
// main.ts:10269
sshConnections.set(scope, { ssh, fingerprint, ownershipId, localPort, remotePort, pid, host, ... })
```

`grep 'sshConnections.set'` over `main.ts` returns **one** hit (line 10269). No writer anywhere stores a connection under a `backendScopeKey`.

**The reader has two branches, and both can compute a composite key.** `main.ts:9918`:

```ts
function activeSshTerminalTarget(webContentsId?: number) {
  const windowRoute = typeof webContentsId === 'number' ? windowConnectionRoutes.get(webContentsId) : null

  if (windowRoute?.registryScoped && windowRoute.connectionId) {          // <-- branch A
    const scope = registrySshScopeForWindowRoute(windowRoute, readDesktopConnectionsRegistry())
    if (!scope) return null
    const state = sshConnections.get(scope)
    return state && state.ssh ? { ssh: state.ssh, scope } : 'pending'     // main.ts:9930 -> always 'pending'
  }

  const profile = windowRoute?.profile ?? primaryProfileKey()
  ...
  const scope = route.connectionId                                        // <-- branch B (#95472)
    ? backendScopeKey(route.connectionId, profile)
    : sshScopeKey(route.source === 'profile' ? profile : null)
  const state = sshConnections.get(scope)
  return state && state.ssh ? { ssh: state.ssh, scope } : 'pending'       // main.ts:9956
}
```

Branch A resolves through `registrySshScopeForWindowRoute` in `apps/desktop/electron/window-connection-route.ts:26`, which ends at line 40:

```ts
return backendScopeKey(route.connectionId, route.profile)   // -> "conn:<conn-id>::default"
```

So the pool is written under `"default"` and read under `"conn:<conn-id>::default"`. `Map.get` misses, the function returns `'pending'`.

Windows become registry-scoped at `main.ts:14198` — `hermes:connection:for` returns `{ ...connection, connectionId: id, registryScoped: true }` — which is the normal path for a `launchMode: "primary"` v2 registry launch, and is stamped onto the window at `main.ts:14207`.

The user-visible throw is `apps/desktop/electron/connection-apply.ts:51`, after `resolveTerminalConnection()` retries once through `ensureBackend()`:

```ts
async function resolveTerminalConnection(getTarget, ensureBackend) {
  let target = getTarget()
  if (target !== 'pending') return target
  await ensureBackend()
  target = getTarget()
  if (target === 'pending') {
    throw new Error('Remote connection is not ready yet. Try again in a moment.')   // line 51
  }
  return target
}
```

### Why chat works but the terminal does not

Chat/profile/session traffic resolves the backend through `ensureRegistryBackend()` / `resolveDesktopRemoteRoute`, which does not consult the `sshConnections` pool by composite key — it uses the live forwarded port. Only the terminal pane gates on `sshConnections.get(scope)`. That is why `desktop.log` shows a completely healthy tunnel at the same instant the pane refuses to open:

```
[ssh-lifecycle] remote hermes version: Hermes Agent v0.20.6 (2026.8.27)
[ssh] forwarding 127.0.0.1:63683 -> 127.0.0.1:36847
[ssh-lifecycle] reusing remote dashboard pid=<PID> port=36847
[boot] Remote Hermes backend is ready
```

Nothing is logged at failure time, because the throw happens in the Electron main process before any RPC is dispatched — the remote host's `gui.log` and `errors.log` stay empty. That makes this bug invisible from the backend side.

### Side effect: remote backend leak

Each failed attempt calls `ensureTerminalBackend()`, which dials `ensureRegistryBackend()` and spawns a **new** remote `hermes serve --isolated` on the SSH host. `ensureTerminalBackend` is already claim-guarded (`main.ts:9962-9973`), so *concurrent* dials for the same scope coalesce — but the claim only covers in-flight dials, and the retry can never succeed, so each fresh click of `+` still leaks one backend.

Observed: **1 → 4** remote backends in ~6 minutes of clicking. Confirmed on the remote host:

```
$ ps -eo pid,etime,rss,cmd | grep '[h]ermes serve'
2356965  13:25:21  280524  .../python3 .../hermes serve --isolated --host 127.0.0.1 --port 0 --ssh-session-token-file .../<redacted>.token --ssh-owner-nonce <redacted>
2381878  13:24:20  622096  .../python3 .../hermes serve --isolated --host 127.0.0.1 --port 0 --ssh-session-token-file .../<redacted>.token --ssh-owner-nonce <redacted>
   2606     23:46  159072  .../python3 .../hermes serve --isolated --host 127.0.0.1 --port 0 --ssh-session-token-file .../<redacted>.token --ssh-owner-nonce <redacted>
```

Each carries its own session-token file and owner nonce, so they do not coalesce. Only backends with zero ESTABLISHED connections are safe to kill.

### Steps to reproduce

1. Configure the desktop with a **v2 registry** `connections.json` (`version: 2`) holding an `ssh` connection, set as `primary`, with `launchMode: "primary"`.
2. Launch the desktop so the window is registry-scoped (`hermes:connection:for` → `registryScoped: true`).
3. Wait for `[boot] Remote Hermes backend is ready` in `desktop.log`; confirm chat and profiles work.
4. Open the terminal pane.

**Expected:** a remote PTY on the SSH host.
**Actual:** `Remote connection is not ready yet. Try again in a moment.`, instantly, every time — plus one leaked remote backend per attempt.

### Relationship to other issues

**This is not #90922.** #90922 was "terminal spawns a LOCAL shell when the SSH connection lives in the v2 registry" — the wrong shell *did* start. Here nothing spawns at all; the gate throws first. #90922 was closed by #95081 (`71d804d00`), and that fix is present in this build and works: the local-shell fallback is gone. This issue is a **regression introduced by that same PR**, which moved terminal routing onto the v2 registry key without moving the pool writer.

**This is not #93888.** Registry drift was ruled out directly: `connection.json` is `"mode": "ssh"`, `connections.json` is `version: 2` with `primary: "<conn-id>"` and a well-formed `ssh` entry. Both files are internally consistent.

**This is adjacent to #95472 / PR #95473, but not covered by them.** #95472 describes **branch B** — v1 settings/profile SSH picking up a registry `connectionId` via `resolveDesktopRemoteRoute` and then keying with `backendScopeKey`. PR #95473 fixes exactly that, and its description states: *"The v2 registry-scoped branch at the top of `activeSshTerminalTarget` is unchanged."*

That branch is **branch A**, which is the one firing here. So #95473, if merged as written, does not fix this. The discriminator is provable statically rather than by assertion: `v1SshTerminalPoolKey` (the helper #95473 introduces) is reached only after the `windowRoute?.registryScoped && windowRoute.connectionId` early-return at line 9921. A registry-scoped window returns at line 9930 and never reaches it.

The deeper invariant is that **no writer anywhere in `main.ts` stores under a `backendScopeKey`** — one `sshConnections.set`, keyed `sshScopeKey(profile)`. Fixing branch B alone leaves branch A reading a key that is never written. Both branches need to agree with the writer, or the writer needs to publish under both keys.

**This is not fixed by PR #96716 either.** #96716 claim-guards three remaining unguarded backend dial sites (`freshGatewayWsUrl`, `requestJsonForProfile`, and the `createRegistryGatewayWsUrlHandler` dependency) so concurrent dials for the same scope coalesce; it changes no pool key and touches neither `activeSshTerminalTarget` nor `bootstrapSshConnectionInner`'s writer. If it merges it will reduce how fast the backend leak below accumulates under racing dials, but the terminal pane will still return `'pending'` on the first attempt and every attempt after it.

Filed separately rather than as a comment on #95472 because the reproduction, the affected code path, and the fix site all differ, and because #95472 is authored around a v1-settings-SSH configuration that this setup does not use.

### What does not fix it

All verified, none change which key the writer uses:

- Save-and-reconnect in Settings → Gateways
- New terminal tab / new window
- Tray **Quit** + relaunch, and `taskkill /F /IM Hermes.exe` + relaunch
- Killing the leaked remote backends on the SSH host
- Updating the client — `origin/main` today is byte-identical at `main.ts:9918-9957`, `window-connection-route.ts:40`, and `main.ts:10158/10269`

A stale-renderer-chunk theory was also investigated and disproved: all 325 dist assets hash-consistent after a full restart, error unchanged.

### Suggested fix

Make the reader agree with the writer, in either direction:

1. **Narrow fix**, symmetric with #95473 — have branch A fall back to the bootstrap key when the composite key misses:
   ```ts
   const state = sshConnections.get(scope) ?? sshConnections.get(sshScopeKey(profile))
   ```
2. **Structural fix** — have `bootstrapSshConnectionInner` publish under the composite key when `metadata.registryConnectionId` is set (it is already in scope at `main.ts:10200`), so registry-scoped tunnels are genuinely per-connection as #95081 intended. This is the option that makes the multi-connection goal of #95081 actually hold.

A regression test asserting *"every key `activeSshTerminalTarget` can return is a key some writer can produce"* would cover both branches and prevent the next recurrence.

### Redaction note

LAN addresses, PIDs, ports, timestamps and version strings are retained deliberately — they carry the diagnosis. SSH session-token filenames and owner nonces are redacted (they authenticate a backend), as are hostnames and account names.

liuhao1024:
## What does this PR do?

After #95081, a v2 registry SSH connection's terminal pane fails instantly with `Remote connection is not ready yet. Try again in a moment.` on every tab and relaunch (#97345).

Root cause: `activeSshTerminalTarget()`'s registry-scoped branch looks the tunnel up under `backendScopeKey(connectionId, profile)`, but the pool's single writer (`bootstrapSshConnectionInner`) publishes every tunnel under its per-profile bootstrap key (`sshScopeKey(profile)`), stamping the entry with `registryConnectionId` instead. The composite lookup can never hit, so the pane is permanently `'pending'`.

This PR makes the registry-scoped branch fall back to the entry the writer stamped with the window's connection id, returning the writer's actual scope. It is the same identity match `managedSshScopeRole` already applies to pool entries (`state.registryConnectionId === connectionId`), and it mirrors the reader-aligns-with-writer direction PR #95473 takes for the v1 branch — the two branches are disjoint (the registry-scoped branch returns before the v1 path is reached), so this does not overlap with #95473.

Deliberately not done here: the structural writer-side option from the issue (publishing registry tunnels under the composite key, which is what would make #95081's multi-connection goal fully hold). That changes the key every per-profile reader, teardown path, and the managed-update drain would see, so it needs a broader consumer audit than this reader-side fix. The fallback below fixes the single-primary-connection regression (#97345's reported setup) without touching pool semantics: with multiple registry connections under one profile the pool still holds one entry, and a mismatched id still reports `'pending'` rather than borrowing another connection's tunnel.

## Related Issue

Fixes #97345

## Type of Change

## Changes Made

- `apps/desktop/electron/window-connection-route.ts`: new `registrySshPoolScopeByConnectionId()` — scans pool entries for the one stamped with the given registry connection id and carrying a live `ssh`, returning the writer's actual scope key.
- `apps/desktop/electron/main.ts`: `activeSshTerminalTarget()`'s registry-scoped branch keeps the composite-key lookup first; on a miss it resolves via `registrySshPoolScopeByConnectionId()` and returns the bootstrap scope so downstream `getSshConnectionState(sshTarget.scope)`, preview-reach scoping, and `sshScope` teardown bookkeeping all hit the real entry.
- `apps/desktop/electron/window-connection-route.test.ts`: tests for the recovery (bootstrap key recovered for a stamped entry) and the negatives (other connection id, unlabelled legacy entry, torn-down entry without `ssh` all stay unmatched).

## How to Test

1. `cd apps/desktop && ../../node_modules/.bin/vitest run electron/window-connection-route.test.ts` — Observed result: 9/9 pass, including the two new tests.
2. `../../node_modules/.bin/tsc -p tsconfig.electron.json --noEmit` — Observed result: passes.
3. `../../node_modules/.bin/vitest run electron/desktop-remote-route.test.ts` — Observed result: 21/21 pass (adjacent #95473 domain unaffected).
4. On a v2 registry `connections.json` (`version: 2`, primary ssh connection, `launchMode: "primary"`) desktop build, opening the terminal pane should start a remote PTY instead of failing with "Remote connection is not ready yet" (per #97345's repro steps). I could not run the full Windows-client → Linux-remote setup locally, so the behavior change is pinned by the unit tests on the scope-resolution logic.

### Code

- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(desktop): ...`)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate (#95473 fixes the v1 branch only; its description explicitly leaves the registry-scoped branch unchanged)
- [x] I've run `pytest tests/ -q` and all tests pass — N/A for this Electron/desktop-only change; ran `vitest run` on the touched project plus `tsc -p tsconfig.electron.json --noEmit` instead (both pass)
- [x] I've tested on my platform: macOS 15.4 (arm64); the reported setup is Windows 11 → Linux SSH — the scope-resolution logic is platform-independent and covered by unit tests

- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — pure key-resolution logic in the Electron main process, no platform APIs touched

### DIFF
--- a/apps/desktop/electron/main.ts
+++ b/apps/desktop/electron/main.ts
@@ -391,7 +391,11 @@ import {
 import { fetchMarketplaceThemes, searchMarketplaceThemes } from './vscode-marketplace'
 import { createWakeIndicatorWindowController } from './wake-indicator-window'
 import { enumerateWindowsFrontToBack, enumerationFailed, readWindowBelow } from './window-below'
-import { registrySshScopeForWindowRoute, WindowConnectionRouteRegistry } from './window-connection-route'
+import {
+  registrySshPoolScopeByConnectionId,
+  registrySshScopeForWindowRoute,
+  WindowConnectionRouteRegistry
+} from './window-connection-route'
 import { installWindowRendererLifecycle } from './window-renderer-lifecycle'
 import { createWindowRevealController } from './window-reveal'
 import {
@@ -9927,7 +9931,18 @@ function activeSshTerminalTarget(webContentsId?: number) {
 
     const state = sshConnections.get(scope)
 
-    return state && state.ssh ? { ssh: state.ssh, scope } : 'pending'
+    if (state && state.ssh) {
+      return { ssh: state.ssh, scope }
+    }
+
+    // The pool's single writer publishes under the per-profile bootstrap key
+    // while stamping the entry with its registry connection id (#97345), so a
+    // composite-key miss must still resolve the live tunnel by that identity
+    // instead of reporting 'pending' forever.
+    const pooledScope = registrySshPoolScopeByConnectionId(sshConnections, windowRoute.connectionId)
+    const pooledState = pooledScope === null ? null : sshConnections.get(pooledScope)
+
+    return pooledState && pooledState.ssh ? { ssh: pooledState.ssh, scope: pooledScope } : 'pending'
   }
 
   const profile = windowRoute?.profile ?? primaryProfileKey()
--- a/apps/desktop/electron/window-connection-route.test.ts
+++ b/apps/desktop/electron/window-connection-route.test.ts
@@ -4,6 +4,7 @@ import { test } from 'vitest'
 
 import {
   normalizeWindowConnectionRoute,
+  registrySshPoolScopeByConnectionId,
   registrySshScopeForWindowRoute,
   WindowConnectionRouteRegistry
 } from './window-connection-route'
@@ -119,4 +120,21 @@ test('uses the canonical default profile scope when a registry SSH route has no 
     ),
     'conn:source-b::default'
   )
+})
+
+test('recovers the bootstrap pool key a registry SSH tunnel was published under', () => {
+  const pool = new Map<string, any>([['', { registryConnectionId: 'source-b', ssh: { alive: true } }]])
+
+  assert.equal(registrySshPoolScopeByConnectionId(pool, 'source-b'), '')
+})
+
+test('does not match another connection, an unlabelled entry, or a torn-down tunnel', () => {
+  const pool = new Map<string, any>([
+    ['research', { registryConnectionId: 'source-a', ssh: { alive: true } }],
+    ['', { registryConnectionId: '', ssh: { alive: true } }],
+    ['worker', { registryConnectionId: 'source-b' }]
+  ])
+
+  assert.equal(registrySshPoolScopeByConnectionId(pool, 'source-b'), null)
+  assert.equal(registrySshPoolScopeByConnectionId(pool, 'source-c'), null)
 })
--- a/apps/desktop/electron/window-connection-route.ts
+++ b/apps/desktop/electron/window-connection-route.ts
@@ -40,6 +40,30 @@ export function registrySshScopeForWindowRoute(
   return backendScopeKey(route.connectionId, route.profile)
 }
 
+export interface RegistrySshPoolEntry {
+  registryConnectionId?: null | string
+  ssh?: unknown
+}
+
+// The sshConnections pool has a single writer that publishes every tunnel under
+// its per-profile bootstrap key while stamping the entry with the registry
+// connection that owns it. A registry-scoped lookup under the composite
+// backendScopeKey can therefore miss a live tunnel; this recovers the writer's
+// actual key from the stamped identity, the same match managedSshScopeRole
+// applies to pool entries.
+export function registrySshPoolScopeByConnectionId(
+  entries: Iterable<readonly [string, RegistrySshPoolEntry | undefined]>,
+  connectionId: string
+): null | string {
+  for (const [scope, entry] of entries) {
+    if (entry?.ssh && entry.registryConnectionId === connectionId) {
+      return scope
+    }
+  }
+
+  return null
+}
+
 export class WindowConnectionRouteRegistry {
   private readonly routes = new Map<number, WindowConnectionRoute>()
 


Enough1122:

**Overall:** Fixes remote SSH terminal routing where v2 registry composite-key miss left live tunnel reported as `pending` forever.

**What it does**
- `apps/desktop/electron/window-connection-route.ts` adds `registrySshPoolScopeByConnectionId(sshConnections, connectionId)` helper to resolve pool's single-writer entry (published under per-profile bootstrap key, stamped with registry connection id #97345).
- `apps/desktop/electron/main.ts:activeSshTerminalTarget` after `sshConnections.get(scope)` miss (no entry for composite `profile:connectionId` scope), falls back to pool lookup by `windowRoute.connectionId` and returns ` { ssh, scope: pooledScope }` if found; else `pending`. `connection-config` export updated accordingly.
- `window-connection-route.test.ts` added.

**Non-blocking notes**
- Fallback guarded with `pooledScope === null ? null : get(pooledScope)` — correct null handling.
- Consider debug log when fallback path taken to aid forensics for #97345-class mismatches.

liuhao1024:
Thanks for the review. Agreed a debug log on the pooled-scope fallback would help forensics for #97345-class mismatches; I kept the diff at the routing fix itself — happy to add the trace if maintainers want it.

zmrlk (commented):
One identity-isolation gap remains at exact head `2e6edf3a1385`.

`registrySshPoolScopeByConnectionId()` returns the **first** live entry whose `registryConnectionId` matches. The same registry connection can legitimately have multiple pool entries — one per composite `connection × profile` scope — plus the bare primary entry. On a composite miss, iteration order can therefore borrow a live tunnel for the right connection but the wrong remote profile.

Minimal reproduction of the production helper:

```ts
const pool = new Map([
  ['conn:source-b::research', {
    registryConnectionId: 'source-b',
    primaryRegistryScope: false,
    ssh: { profile: 'research' }
  }],
  ['', {
    registryConnectionId: 'source-b',
    primaryRegistryScope: true,
    ssh: { profile: 'default' }
  }]
])

registrySshPoolScopeByConnectionId(pool, 'source-b')
// actual: 'conn:source-b::research'; expected fallback: ''
```

This distinction already exists in [`managedSshScopeRole()`](https://github.com/NousResearch/hermes-agent/blob/32fe1293244f186c75e82090ed06fe416739f146/apps/desktop/electron/managed-ssh-update.ts#L146-L155): matching-id entries are classified as `primary` only when `primaryRegistryScope === true`; otherwise they are pool entries. The writer also stamps that flag explicitly.

Since this fallback is recovering the `startHermes` primary that the direct composite lookup missed, please require `entry.primaryRegistryScope === true` (or constrain the helper to expected writer scopes/profile) rather than scanning by connection id alone. A regression test should insert a same-id non-primary composite entry before the primary and prove the primary wins; same-id non-primary-only should remain unmatched.

The new Windows reproduction on #97345 strongly confirms the reported single-primary failure and the direction of this PR. This guard is about not fixing that case by weakening profile isolation.

kshitijk4poor:
Landed on `main` via #103654 as `2afd17a9d7`, with your authorship preserved on the fix commit, @liuhao1024 — thank you. Closing in favour of the merged salvage.

PR #95992, opened 2026-08-27
fix(desktop): navigate to branched session route

mashenchina-max:
## Summary

When branching the session already open in the primary workspace, navigate the router to the child session before resuming it.

This is a narrow follow-up to #93663. That change correctly made the child the selected primary session, but the URL could remain on the parent. The primary chat then saw `routeSessionMismatch` and kept its centered loading spinner mounted even though the branched child existed and was readable.

## Root cause

`forkBranch()` called:

```ts
await resumeSession(routedSessionId)
```

`resumeSession()` updates `$selectedStoredSessionId`, but it does not change the route. The resulting state was:

- routed session: parent
- selected session: child
- `routeSessionMismatch`: true

This also let route-resume keep trying to restore the parent instead of converging on the child.

## Change

- Navigate to `sessionRoute(routedSessionId)` with `replace: true` before resuming the child.
- Add a regression assertion to the existing primary-workspace branch test.
- Preserve the background/sidebar branch path unchanged, so branching another session still opens a tile and does not steal focus (#69750).

## Regression proof

With the production `navigate(...)` line temporarily removed, the focused test fails with:

```text
expected navigate to be called with ['/branch-stored', { replace: true }]
Number of calls: 0
```

Restoring the line makes the complete `use-session-actions.test.tsx` suite pass: **83/83**.

## Verification

- `npx vitest run --project ui src/app/session/hooks/use-session-actions.test.tsx` — 83 passed
- `npm run check` — passed (typecheck, lint, UI, Desktop, platform, and plugin gates; Electron platform tail: 586/586)
- Packaged Desktop manual verification on Linux: branching the current session opened the child immediately; no `Session not found` retry-loop log appeared after waiting beyond the previous retry interval.

## Relationship to existing reports

- Follow-up to #93663 / #93444: completes the main-workspace transition by keeping route and selection aligned.
- Related to #93959, specifically the variant where the child is already durable/readable but the Desktop remains on the loading spinner.
- Does not overlap #94208's gateway persistence fix for seeded branches whose child row does not yet exist.

PR #95179, opened 2026-08-26
feat(desktop): projects own their gateway; sessions inherit

valvesss:
## Summary

Gateway belongs on the **project**, not on New session.

When you create a project you pick which registered host it lives on, browse that host’s filesystem (without re-homing the desktop chrome), and run `projects.create` there. Sessions started with `+` inside that project inherit the same gateway via `startSessionOnSource`. Local projects stay unlabeled; remote ones show a quiet origin suffix (Cursor-style).

This supersedes [#94457](https://github.com/NousResearch/hermes-agent/pull/94457). Useful infra from that lineage is preserved; the per-click New session source picker is not.

### Review repair (`f9968bcca3`)

Addresses [@andrexibiza](https://github.com/andrexibiza) reviews ([5026257348](https://github.com/NousResearch/hermes-agent/pull/95179#pullrequestreview-5026257348), [5026278580](https://github.com/NousResearch/hermes-agent/pull/95179#pullrequestreview-5026278580)):

1. **Qualified project address** — `ProjectAddress = { connectionId, profile, backendProjectId }`. Stamped on tree/create; overlay/scope migrate bare ids; enter/fetch/CRUD/`+` consume the address (RPC still sends bare backend id). Witnesses: same id on two gateways; same connection + same id in two profiles; same auto path local+SSH; refresh cannot transfer ownership.
2. **Reciprocal local** — foreign merge includes every non-active source (including `local` when chrome is remote); `+` routes through `startSessionOnSource` with the project’s stored profile.
3. **Desktop token mode** — `should_require_dashboard_auth()` is a pure exposure decision again. `start_server` admits token mode only with SSH one-shot token+nonce or local Desktop parent-ownership proof. Ambient `HERMES_DESKTOP=1` alone fails closed under a non-loopback public URL (also: missing token, non-headless, non-loopback bind).

Also coordinating with open [#95045](https://github.com/NousResearch/hermes-agent/pull/95045).

### 1. Projects overview — remote hint vs local

Projects overview with remote mimir hints and unlabeled local projects

### 2. Inline expand

Brokk project inline expand with mimir hint and session previews

### 3. Entered project

Entered Brokk project view with mimir hint in header and branch row

## Test plan

- [x] Same `backendProjectId` on two gateways: enter/fetch/update/delete hit the clicked connection only
- [x] Same auto-project path on local + SSH: address keys differ; cwd tie prefers chrome
- [x] Same connection, same project ID, two profiles: fully isolated
[...]

PR #89736, opened 2026-08-19
fix(desktop): preserve project session profiles

honki12345:
## What does this PR do?

Fixes Hermes Desktop's **Filters → Show → Profile** option for session rows rendered inside **Project** grouping.

Project overview and drill-in rows come from the shared `projects.tree` projection. That projection dropped the persisted `profile_name`, so the renderer received no `session.profile`, normalized the row to `default`, and intentionally suppressed the profile glyph. The session itself was still correctly owned by the named profile; only its display metadata was missing.

This PR projects the existing profile identity into the response once, at the shared backend boundary. Named-profile project rows can therefore use the existing `ProfileTag` rendering path, while rows whose persisted profile is `NULL` retain the canonical `default` identity and existing default-profile UI behavior.

This is distinct from #86258 / #86259, which covers ordinary flat **All Profiles** recents and explicitly leaves scoped/project views unchanged.

## Related Issue

Fixes #89731

Related: #72767, #86258

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `tui_gateway/server.py`
  - preserve the session row's owning profile in the shared project-tree projection
  - normalize a missing persisted profile to the canonical `default` identity
  - cover both `projects.tree` overview rows and `projects.project_sessions` drill-in rows through their shared projection
- `tests/tui_gateway/test_projects_rpc.py`
  - exercise the real temporary `state.db` → project RPC response path
  - verify named-profile and default-profile identities in both overview and drill-in payloads

## How to Test

1. Select a named Desktop profile and create a session inside a project/worktree.
2. Set the session sidebar to **Grouping → Project**.
3. Enable **Show → Profile**.
4. Confirm the project session row shows the existing named-profile glyph and `Profile: <name>` accessible label.
5. Confirm default-profile rows retain their existing default-profile behavior.

Automated verification:

```text
RED (before production change):
  scripts/run_tests.sh tests/tui_gateway/test_projects_rpc.py \
    -k project_session_rows_carry_profile_identity -q
  2 failed with KeyError: 'profile'

GREEN (after production change):
  same focused command
  2 passed

Related project RPC/tree suite:
[...]

PR #97359, opened 2026-08-28
fix(desktop): route project session branches by owner

evan-bradford:
## Summary

Create a Desktop conversation branch through the parent session's owning profile when that parent comes from a profile-scoped Project tree.

This fixes the observed case where a session owned by a non-default profile was visible under a Project, but **Branch** could not resolve/read it or created the child through the active/default profile.

## Root cause

Profile ownership was lost across three boundaries:

1. Profile-scoped `projects.tree` / `projects.project_sessions` rows were not self-describing.
2. Sidebar and tile branch callbacks forwarded only the session ID.
3. `branchStoredSession()` could let an ownerless Recents duplicate mask the authoritative Project-tree row, and an explicitly forwarded profile skipped or mis-scoped parent metadata resolution (losing the worktree `cwd` or borrowing it from a same-ID session in another profile).

## Fix

- Stamp the authoritative requested profile onto project-tree preview and nested session rows after scoped DB construction.
- Include project-tree sessions in stored-session owner resolution.
- Prefer a profile-bearing row over an ownerless duplicate for the same stored ID.
- Forward optional owner profile through sidebar and tile branch actions.
- Route parent message reads and `session.create` through that profile.
- Resolve parent metadata inside the explicitly forwarded profile, preserving the correct worktree `cwd` and failing closed rather than borrowing another profile's metadata.
- Preserve profile-less/single-profile behavior by omitting profile routing when no owner exists.

## Regression coverage

- Parent exists only in a non-default profile Project tree.
- Ownerless Recents duplicate cannot mask an owned Project-tree row.
- Explicit owner profile plus uncached parent still resolves worktree `cwd`.
- Same stored ID in two profiles cannot route to one profile while inheriting the other's `cwd`.
- Genuinely profile-less parent remains unscoped.
- Sidebar wiring and session-tile delegation carry ownership.
- Profile-scoped project overview and drill-in rows carry the requested profile.
- Native Electron E2E: visible persisted parent row → Session actions → Branch → nested child.

## Verification on current `main`

- Desktop targeted Vitest: **123 passed**
- Python profile/project suites: **39 passed**
- Desktop typecheck: passed
- Touched-file ESLint: passed
- Production Desktop build + `assert-dist-built`: passed
- Added-line security scan: no hardcoded-secret, shell-injection, eval/exec, or pickle findings
- Initial independent review caught same-ID cross-profile metadata leakage; the exact default-vs-work CWD case was reproduced RED and is GREEN in follow-up `2fc2408fc`
- Independent filesystem-aware runtime-head review of `2fc2408fc` — **PASS, no blockers**
- Test-only tip `b5f9c2619` additionally pins explicit-profile fetch failure against a foreign same-ID cache row
- Native macOS E2E on exact runtime head `2fc2408fc`: arm64 production build, then visible persisted parent → Session actions → Branch → nested child — **1/1 passed in 11.1s**

## Relationship to open PRs

- **#89736** fixes profile badge metadata by projecting persisted `profile_name` (defaulting missing values). This PR fixes branch routing and stamps the *requested scoped profile*, which remains authoritative for legacy rows whose persisted `profile_name` is null. The behavioral scope is additive, though the response field may conflict mechanically if #89736 merges first.
- **#95992** fixes URL navigation after a child is already created. This PR fixes owner routing for parent lookup/read/create before that point. The fixes are complementary.
- **#95179** is a much broader multi-gateway Project ownership feature; this PR is the narrow existing-session bug fix.

## Cache/profile safety

No prompt, message history, memory, or toolset state is mutated. Ownership is carried as request/display metadata and used only to select the existing profile-scoped session APIs.

### DIFF
--- a/apps/desktop/e2e/session-branch.spec.ts
+++ b/apps/desktop/e2e/session-branch.spec.ts
@@ -1,0 +1,75 @@
+import {
+  buildAppEnv,
+  createSandbox,
+  launchDesktop,
+  type MockBackendFixture,
+  waitForAppReady,
+  writeEnvFile,
+  writeMockProviderConfig
+} from './fixtures'
+import { startMockServer } from './mock-server'
+import { RealSessionBuilder } from './real-session-builder'
+import { expect, test } from './test'
+
+const SESSION_LABEL = 'E2E persisted conversation branch parent'
+
+async function setupSeededBackend(): Promise<MockBackendFixture> {
+  const mock = await startMockServer()
+  const sandbox = createSandbox('session-branch')
+
+  writeMockProviderConfig(sandbox.hermesHome, mock.url)
+  writeEnvFile(sandbox.hermesHome)
+
+  const builder = await RealSessionBuilder.start(sandbox.hermesHome)
+
+  try {
+    await builder.createSession({ title: SESSION_LABEL, turns: [SESSION_LABEL] })
+  } finally {
+    await builder.close()
+  }
+
+  const { app, page } = await launchDesktop(buildAppEnv(sandbox))
+
+  return {
+    app,
+    page,
+    mock,
+    mockUrl: mock.url,
+    sandbox,
+    cleanup: async () => {
+      await app.close()
+      await mock.close()
+      sandbox.cleanup()
+    }
+  }
+}
+
+test.describe('persisted session branching', () => {
+  let fixture: MockBackendFixture
+
+  test.beforeAll(async () => {
+    fixture = await setupSeededBackend()
+    await waitForAppReady(fixture, 120_000)
+  })
+
+  test.afterAll(async () => {
+    await fixture?.cleanup()
+  })
+
+  test('branches from the visible session-row action and renders the nested child', async () => {
+    const parentLabel = fixture.page.getByText(SESSION_LABEL, { exact: true }).first()
+
+    await expect(parentLabel).toBeVisible({ timeout: 30_000 })
+
+    const parentRow = parentLabel.locator('xpath=ancestor::div[contains(@class, "row-hover")]').first()
+
+    await parentRow.hover()
+    await parentRow.getByRole('button', { name: 'Session actions' }).click()
+    await fixture.page.getByRole('menuitem', { name: 'Branch', exact: true }).click()
+
+    const childLabel = fixture.page.getByText(/draft: branch #\d+/i).first()
+
+    await expect(childLabel).toBeVisible({ timeout: 30_000 })
+    await expect(parentLabel).toBeVisible()
+  })
+})
--- a/apps/desktop/src/app/chat/sidebar/index.tsx
+++ b/apps/desktop/src/app/chat/sidebar/index.tsx
@@ -299,7 +299,7 @@ interface ChatSidebarProps extends React.ComponentProps<typeof Sidebar> {
   onResumeSession: (sessionId: string, session?: SessionInfo) => void
   onDeleteSession: (sessionId: string) => void
   onArchiveSession: (sessionId: string) => void
-  onBranchSession: (sessionId: string) => void
+  onBranchSession: (sessionId: string, profile?: string) => void
   onNewSessionInWorkspace: (path: null | string) => void
   /** Create a brand-new session and open it as a tile on `dir`. */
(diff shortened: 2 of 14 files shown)


evan-bradford:
## Live macOS deployment verification — 2026-08-28

I built the combined reviewed source at `96d0deb87d8726ed784e83595177bf2a42c70b1b` natively on Apple Silicon and installed it atomically on the real Desktop client.

Provenance and packaging:

- host/target: `Darwin arm64`
- executable: `Mach-O 64-bit executable arm64`
- bundle: `com.nousresearch.hermes`, package type `APPL`
- strict ad-hoc signature verification: passed
- installed `install-stamp.json`: exact commit `96d0deb87d8726ed784e83595177bf2a42c70b1b`
- isolated native Electron branch E2E on the same combined source: **1/1 passed in 13.9s**

After backend restart and app relaunch, the actual affected parent session appears exactly once through the `sfb` project drill-in, carries `profile: "sfb"`, and carries its correct linked-worktree CWD. The installed app is connected to the restarted backend with a matching current token and one visible native window.

I intentionally did not create a throwaway branch in the production session DB; the exact branch interaction is exercised by the native Electron E2E, while live verification remained read-only.

evan-bradford:
## Follow-up: keep launch-profile parent metadata in branch owner path (`51b6f2fe`)

This PR's explicit-profile arm went fail-closed for the **launch profile** — a regression exposed in rollout (default-profile branches appeared to fail; the branch was actually created but with no workspace until a tree/session refresh repainted it).

### Root cause
`stamp_profile` now stamps default-profile project-tree rows `profile="default"` (default IS the launch home, so `_response_profile_name` falls to `_current_profile_name()`). `branchStoredSession` then forwards an explicit `"default"` into `resolveStoredSession(id, "default")`, whose fail-closed string-arm (`utils.ts:1348-1351`) returns `undefined` on a by-id `getSession` miss — dropping the parent's `cwd` before `forkBranch`. The branch lands with no workspace → nothing appears where the click happened → navigate-away-and-back (refetch) "fixes" it.

### Fix
Soft-fall to the cached row **only** when the explicit profile equals the active gateway profile — that is not a cross-profile claim, so no metadata can leak. Genuinely foreign profiles still fail closed (sfb routing unchanged; the existing `'work'` foreign-profile tests still pass).

### Verification
- New regression test (launch-profile by-id miss) was **RED** before the fix (asserted `cwd`/parent survive — failed with cwd dropped).
- **GREEN** after: this worktree's targeted suite `use-session-actions.test.tsx` + `resolve-stored-session.test.ts` → **100/100 passed**.
- eslint clean on both files; change is 47 lines in 2 files.

The launch-row stamp corner was previously flagged as untested (Opus review finding #5 on the cross-profile branch fix) — this closes that gap on the branch side.

evan-bradford:
Owner note for any agent resuming this PR: commit `51b6f2fe3` was authored and pushed today (Aug 28) from the 'Fix branch creation in sfb profile' branch session's child, coordinating the default-profile launch-row regression follow-up. Treat it as owned work — not a conflict. Diagnostics, the RED-GREEN regression test, and the full verification (100/100 targeted tests, lint clean) are in this thread's previous comment. Do not rebase it away; if the branch needs force-pushing for any reason, preserve or re-cherry-pick `51b6f2fe3` on top.

OutThisLife:
Superseded by #100879, which consolidates the full branch-load path.\n\nCarried from here: the gateway-side profile stamping on scoped project rows (), committed under your authorship, plus preferring a self-describing row over an ownerless Recents duplicate. The routing half is subsumed by full  owner routing \u2014 a profile name alone can't disambiguate two connections that both expose , which is the case in the original report. The owner now resolves inside , so the sidebar/tile profile passthrough isn't needed.

OutThisLife:
Superseded by #100879, which consolidates the full branch-load path.

Carried from here: the gateway-side profile stamping on scoped project rows (`stamp_profile`), committed under your authorship, plus preferring a self-describing row over an ownerless Recents duplicate.

The routing half is subsumed by full `(connection, profile)` owner routing — a profile name alone can't disambiguate two connections that both expose `default`, which is the case in the original report. The owner now resolves inside `branchStoredSession`, so the sidebar/tile profile passthrough isn't needed.

PR #92828, opened 2026-08-23
fix(dashboard-auth): treat non-JWT tokens as 'not my token', not IDP outage

mpartipilo:
## Problem

The `self-hosted` and `nous` dashboard-auth OIDC providers wrap `get_signing_key_from_jwt()` in a catch-all that maps **any** exception to `ProviderError("JWKS lookup failed: ...")`. The middleware translates `ProviderError` into HTTP **503 `{"detail": "Auth provider 'self-hosted' unreachable"}`**.

But PyJWT raises `jwt.DecodeError` (an `InvalidTokenError`, **not** a `PyJWKClientError`) when the token isn't a JWT at all — e.g. a session minted by a *different* provider, or a stale cookie. So a foreign/opaque token in the session cookie is misreported as an IDP outage, and the user sees a misleading 503 instead of a clean fall-through to the next provider / re-login.

Observed in the wild (self-hosted Authelia OIDC + `basic` password provider both registered):

```
dashboard-auth: provider 'self-hosted' unreachable during verify: JWKS lookup failed: DecodeError('Not enough segments')
```

## Fix

Catch `jwt.InvalidTokenError` separately in `_verify_id_token` (self_hosted) and `_verify_jwt` (nous) and raise `InvalidCodeError`, which `verify_session` already maps to `None` — the "not my token" contract. `PyJWKClientError` (a genuine JWKS fetch failure) still maps to `ProviderError` → 503, preserving the real-outage path.

## Tests

Added `test_foreign_opaque_token_returns_none` to both provider test suites, asserting a `jwt.DecodeError` from the JWKS client yields `None` (not `ProviderError`). Verified RED→GREEN: both tests fail on unpatched source with the exact `ProviderError: JWKS lookup failed: DecodeError('Not enough segments')`, and pass with the fix. Full suites: 65 passed.

PR #97362, opened 2026-08-28
fix(dashboard-auth): treat an unrecognised token as None, not as an unreachable IDP

### LINKED ISSUE #97361: [Bug]: a foreign or stale session token is reported as an unreachable IDP (503), so clients can never recover
### Bug Description

Both bundled JWT dashboard-auth providers (`plugins/dashboard_auth/nous`, `plugins/dashboard_auth/self_hosted`) map **every** exception out of `get_signing_key_from_jwt` to `ProviderError`:

```python
except jwt.PyJWKClientError as exc:
    raise ProviderError(f"JWKS lookup failed: {exc}") from exc
except Exception as exc:  # pragma: no cover - defensive
    raise ProviderError(f"JWKS lookup failed: {exc!r}") from exc
```

A token the provider did not mint lands in that branch. PyJWT cannot split an opaque session string into segments, so it raises `DecodeError("Not enough segments")` locally, **before any network call**. The provider contract says the opposite (`base.py`): return `None` for a token you do not recognise, raise `ProviderError` only when the IDP is unreachable.

`gated_auth_middleware` turns `ProviderError` into **HTTP 503** (`{"detail": "Auth provider 'nous' unreachable"}`). So a cookie minted by another stacked provider, or left over after an operator changed the auth configuration, makes every request 503 instead of 401, and the caller never receives the 401 that would send it back to `/login`.

### Steps to Reproduce

On a gated dashboard reachable over the network:

1. Configure `dashboard.basic_auth`, sign in from a browser and from Hermes Desktop. Both now hold a `hermes_session_at` cookie minted by the `basic` provider (an opaque HMAC blob, not a JWT).
2. Register a Nous Portal dashboard client (`hermes dashboard register`), remove the `basic_auth` block, restart. `/api/status` correctly reports `auth_providers: ["nous"]`.
3. Reload the dashboard, or relaunch Desktop, **without clearing cookies first**.

Observed: every request answers 503 `Auth provider 'nous' unreachable`, the login page is unreachable (the gate 503s before rendering it), and `logs/dashboard-auth.log` fills with

```json
{"event":"session_verify_failure","provider":"nous","reason":"provider_unreachable"}
```

with the server log line `dashboard-auth: provider 'nous' unreachable during verify: JWKS lookup failed: DecodeError('Not enough segments')`. The Portal is perfectly reachable throughout (`curl https://portal.nousresearch.com/.well-known/jwks.json` returns 200 in ~55 ms from the same container).

Hermes Desktop is the worst case: it reads the 503 as "gateway unavailable" and retries forever ("Desktop boot failed: Could not reach the remote Hermes gateway while refreshing its WebSocket ticket"), with no path in the UI to re-authenticate. The only way out is clearing the app's cookie store.

### Expected Behavior

An unrecognised token is `None`: the middleware tries the next provider and ends on a clean 401, which the SPA and Desktop already handle by returning to `/login`. `ProviderError` (and therefore 503) is reserved for a JWKS endpoint that genuinely could not be fetched, so a transient IDP outage does not log valid users out.

PyJWT already distinguishes these: `PyJWKClientConnectionError` for fetch failures, `PyJWKClientError` for "no JWK matches this kid" (raised only after PyJWT's own one-shot JWK-set refresh), `InvalidTokenError`/`DecodeError` for a token it cannot parse.

### Actual Behavior

Any foreign or stale token is reported as an unreachable IDP, and the 503 masks the 401 that would let the client recover on its own.

### Affected Component

- Dashboard auth (`plugins/dashboard_auth/nous`, `plugins/dashboard_auth/self_hosted`, `hermes_cli/dashboard_auth/middleware.py`)
- Hermes Desktop (symptom surface)

### Environment

Hermes Agent v0.20.6 (2026.8.27), Docker image on a container platform, gated bind behind a TLS-terminating edge proxy, Desktop 0.20.6 on macOS 15.

woprrr:
## What does this PR do?

Both bundled JWT providers mapped every exception out of `get_signing_key_from_jwt` to `ProviderError`, including the purely local `DecodeError` PyJWT raises for a string it cannot parse as a JWT. The provider contract is the opposite: `None` for a token you do not recognise, `ProviderError` only when the IDP cannot be reached.

Because the middleware turns `ProviderError` into 503, a session cookie minted by another stacked provider (or left over after an operator changed the auth configuration) made every request 503 instead of 401, and the caller never received the 401 that sends it back to `/login`. Hermes Desktop retried the "unavailable gateway" forever with no way out of the UI.

Classification now follows what PyJWT actually raises:

| PyJWT exception | Meaning | Now |
|---|---|---|
| `PyJWKClientConnectionError` | JWKS endpoint could not be fetched | `ProviderError` (503), unchanged |
| `InvalidTokenError` / `DecodeError` | not a token we minted | `InvalidCodeError`, which `verify_session` maps to `None` (401) |
| other `PyJWKClientError` | no JWK matches the kid, after PyJWT's own one-shot refresh | `InvalidCodeError` (401) |

The same three lines were wrong in the self-hosted OIDC provider, which the middleware documents as stacking with the nous provider ("a self-hosted-OIDC session hits the `nous` provider first"), so both are fixed together.

## Related Issue

Fixes #97361

## Type of Change

## Changes Made

- `plugins/dashboard_auth/nous/__init__.py`: split the `get_signing_key_from_jwt` handler per the table above.
- `plugins/dashboard_auth/self_hosted/__init__.py`: same split.
- `tests/plugins/dashboard_auth/test_nous_provider.py`, `tests/plugins/dashboard_auth/test_self_hosted_provider.py`: 4 new tests, and 2 existing ones updated (see below).

### Note for review: two existing tests changed

`test_jwks_unreachable_raises` (both files) raised a bare `PyJWKClientError("fetch failed")` to stand in for an unreachable endpoint. PyJWT never raises that for a fetch failure: `fetch_data` raises `PyJWKClientConnectionError`. The tests encoded the very conflation this PR fixes, so they now raise the subclass PyJWT actually emits. The unreachable-IDP behaviour they guard (503, no forced logout) is unchanged and still asserted.

## How to Test

Reproduce the trap:

1. Gate a network-reachable dashboard with `dashboard.basic_auth`, sign in from a browser and from Desktop.
2. Register an OAuth client, remove the `basic_auth` block, restart. `/api/status` reports `auth_providers: ["nous"]`.
3. Reload without clearing cookies.

Before: every request returns 503 `Auth provider 'nous' unreachable`, `/login` is unreachable, `logs/dashboard-auth.log` fills with `provider_unreachable`, and Desktop loops on "Could not reach the remote Hermes gateway while refreshing its WebSocket ticket".

After: the stale cookie yields a clean 401, the browser lands on `/login`, Desktop offers Sign in.

Automated coverage:

- `test_foreign_opaque_token_returns_none` (both providers): the exact production shape, an opaque non-JWT session, must be `None`.
- `test_unknown_signing_key_returns_none` (both providers): a well-formed JWT whose kid matches no JWK is foreign, not an outage.
- `test_jwks_unreachable_raises*` (both providers): a real fetch failure still raises, so an IDP outage never logs a valid user out.

Each fix was verified as a genuine regression by reverting it independently: both new tests fail on `main` with the ProviderError the fix removes.

### Code

- [x] I searched for existing PRs and issues to make sure this isn't a duplicate
- [x] I've run the affected suites with `scripts/run_tests.sh`: `tests/plugins/` plus the dashboard-auth files, 1849 passed against 1845 on `main` at the same commit, with the same 70 pre-existing failures either way
- [x] I've tested on my platform: macOS 15 (Darwin 25.6), Python 3.11, plus the live reproduction on a Linux container running the v2026.8.27 image

- [x] N/A: no config key, no CLI surface, no doc statement changes (the contract this restores is the one `base.py` already documents)
- [x] Cross-platform: pure Python exception handling

### DIFF
--- a/plugins/dashboard_auth/nous/__init__.py
+++ b/plugins/dashboard_auth/nous/__init__.py
@@ -436,8 +436,25 @@ class NousDashboardAuthProvider(DashboardAuthProvider):
             signing_key = self._get_jwks_client().get_signing_key_from_jwt(
                 access_token
             )
-        except jwt.PyJWKClientError as exc:
+        except jwt.exceptions.PyJWKClientConnectionError as exc:
+            # The JWKS endpoint could not be fetched: we can neither confirm
+            # nor deny this token, so the middleware must answer 503 and keep
+            # the session rather than logging a valid user out on an outage.
             raise ProviderError(f"JWKS lookup failed: {exc}") from exc
+        except jwt.InvalidTokenError as exc:
+            # Not something this provider minted. The dashboard stacks
+            # providers, so a session cookie from another one (or left over
+            # from a previous auth configuration) lands here as a string PyJWT
+            # cannot even parse. Contract: unrecognised tokens are None, which
+            # lets the middleware try the next provider and end on a clean 401.
+            # Reporting it as an outage instead strands the caller: it keeps
+            # retrying an "unreachable" gateway and never returns to /login.
+            raise InvalidCodeError(f"not a Nous Portal token: {exc}") from exc
+        except jwt.PyJWKClientError as exc:
+            # No JWK matches the token's kid, and PyJWT already refreshed the
+            # set once before raising. The endpoint answered, so this is a
+            # foreign token, not an outage.
+            raise InvalidCodeError(f"no matching signing key: {exc}") from exc
         except Exception as exc:  # pragma: no cover - defensive
             raise ProviderError(f"JWKS lookup failed: {exc!r}") from exc
 
--- a/plugins/dashboard_auth/self_hosted/__init__.py
+++ b/plugins/dashboard_auth/self_hosted/__init__.py
@@ -617,8 +617,22 @@ class SelfHostedOIDCProvider(DashboardAuthProvider):
             signing_key = self._get_jwks_client().get_signing_key_from_jwt(
                 id_token
             )
-        except jwt.PyJWKClientError as exc:
+        except jwt.exceptions.PyJWKClientConnectionError as exc:
+            # The IDP's JWKS endpoint could not be fetched: neither confirm nor
+            # deny, so the middleware answers 503 instead of logging a valid
+            # user out on a transient outage.
             raise ProviderError(f"JWKS lookup failed: {exc}") from exc
+        except jwt.InvalidTokenError as exc:
+            # Not an ID token this provider issued. Providers stack (the
+            # middleware documents exactly this: a nous session hitting the
+            # self-hosted provider and vice versa), and an unrecognised token
+            # is None per the contract, so the caller gets a clean 401 rather
+            # than a 503 it will keep retrying.
+            raise InvalidCodeError(f"not an ID token from this issuer: {exc}") from exc
+        except jwt.PyJWKClientError as exc:
+            # No JWK matches the token's kid, and PyJWT already refreshed the
+            # set once. The endpoint answered, so this is a foreign token.
+            raise InvalidCodeError(f"no matching signing key: {exc}") from exc
         except Exception as exc:  # pragma: no cover - defensive
             raise ProviderError(f"JWKS lookup failed: {exc!r}") from exc
 
--- a/tests/plugins/dashboard_auth/test_nous_provider.py
+++ b/tests/plugins/dashboard_auth/test_nous_provider.py
@@ -590,14 +590,47 @@ class TestVerifySession:
 
     def test_jwks_unreachable_raises_provider_error(self, provider, rsa_keypair):
         token = _mint_token(rsa_keypair)
-        # Replace the patched client so it raises.
+        # Replace the patched client so it raises. PyJWT raises this subclass
+        # (not the bare PyJWKClientError) when the JWKS endpoint cannot be
+        # fetched, which is the only case where we can neither confirm nor
+        # deny the token and therefore must surface a 503.
         bad_client = MagicMock()
-        bad_client.get_signing_key_from_jwt.side_effect = jwt.PyJWKClientError(
-            "fetch failed"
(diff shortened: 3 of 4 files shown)


woprrr:
Superseded in spirit by #92828, which predates this by five days and which I failed to find before opening this: my duplicate search keyed on the symptom wording rather than on "not my token".

That PR is the better base. It names the condition with a `MalformedTokenError` subtype, it also fixes the token-endpoint paths (a malformed token from the IDP becomes a 400-class error instead of a silent `None`), and its test drives the real `PyJWKClient` instead of a mock.

Rather than have two PRs competing over the same functions, I have proposed the one classification delta this branch adds as a patch on #92828: an unknown `kid` is a foreign token too, since PyJWT already refetches the JWK set once before raising `PyJWKClientError`, while genuine fetch failures raise `PyJWKClientConnectionError`. Maintainers: happy for this PR to be closed in favour of #92828, or to reduce it to that delta as a follow-up once #92828 lands. Whichever keeps the queue clean.

The report in #97361 stands on its own, since it documents the user-visible consequence: the 503 masks the 401, so a desktop client loops on "gateway unavailable" with no recovery path in the UI.

woprrr:
Closing in favour of #92828, which predates this by five days and is the better base.

The one classification this branch added that #92828 did not cover, an unknown `kid` being a foreign token rather than an outage, is now offered as a patch on that author's branch (mpartipilo/hermes-agent#1), so it can land with their PR instead of competing with it. Nothing is lost by closing here.

The report in #97361 stays open on its own: it documents the user-visible consequence (the 503 masks the 401, so a desktop client loops on "gateway unavailable" with no recovery path in the UI), which #92828 fixes.

PR #95860, opened 2026-08-26
fix(honcho): a transient read failure is not an empty credential store

FestoneX:
## What does this PR do?

`plugins/memory/honcho/oauth.py::_persist_credential` docstring promises **"leaving all else intact"** — but it seeds its write from `_read_config()`, which returns `{}` on **any** read failure, and `_atomic_write_config()` replaces the whole file via `os.replace`, which needs only a writable parent. So a present-but-unreadable `honcho.json` does not stop the overwrite: the next persist replaces the store with a single-host block, destroying every other host's credentials and the root config.

Two paths reach it:

1. **Automatic token refresh** (`_rotate_and_persist`, called from `ensure_fresh_token` near expiry) — no user action required. The store can be readable when the token is cached and unreadable by the time the refresh fires (EACCES after a root-owned write, EIO, a stalled mount).
2. **Fresh login** (`install_grant`) — the root-merge is seeded from the same tolerant reader.

This is the defect class of #75206 (P1, fixed for the core auth store in #75258, and the same shape as my open #88822 for the photon plugin): *a failed read must never become an empty store that then gets written back.*

## Related Issue

No issue filed — #75206 is the accepted precedent for the class; #88822 is the photon sibling.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

One file, `plugins/memory/honcho/oauth.py`:

- **`_read_config_strict()`** (new) for the write paths: missing file still bootstraps as `{}`; an unreadable file **raises** with the store untouched; genuine corruption still degrades but **preserves a `.corrupt` copy first** — a truncated store usually holds the other hosts' tokens verbatim, and the next full-file write destroys the only recoverable version. Mirrors `hermes_cli.auth._load_auth_store`.
- **`_rotate_and_persist`** takes the strict read **before** the exchange — rotation is single-use, so an exchange whose result cannot be persisted loses the grant. An unreadable store is now a refresh failure (cooldown set, callers fail open per this module's documented contract), not a reason to write a single-host file. The dict threads through to `_persist_credential`, which also closes the re-read race between the locked read and the persist.
- **`install_grant`** seeds its root-merge from the strict reader; a raise surfaces in the interactive setup flow.
- **Read paths unchanged**: `_read_config` keeps its fail-open contract (unreadable = no credential) and now documents why it must never feed a full-file write. Both readers use `utf-8-sig`, so a BOM'd (perfectly healthy) store is no longer classified as corruption — that variant needed no filesystem fault at all.

[...]

Issue #66125, opened 2026-07-17
honcho: named-profile host blocks don't fall back to the default host's apiKey — inherit or warn?


teknium1:
## Summary

When a named Hermes profile resolves its Honcho host key (e.g. `hermes_coder`) and that host block in honcho.json has no `apiKey`, the resulting `HonchoClientConfig` comes back with `api_key=None` even when the shared honcho.json carries a static key on the default `hermes` host block.

Reproduced E2E on current main (post #62290/#66052):

```
honcho.json:
{
  "defaultHost": "selfhosted",
  "hosts": {
    "hermes":     {"apiKey": "shared-key"},
    "selfhosted": {"baseUrl": "http://192.168.1.50:8000"}
  }
}

active profile: coder → host key "hermes_coder"
HonchoClientConfig.from_global_config() → api_key=None
```

## Design question, not (necessarily) a bug

`resolve_active_host()` documents named-profile isolation as deliberate: named profiles do NOT inherit `defaultHost`, precisely so profiles stay isolated islands. The open question is whether *static API keys* should be treated the same way (isolation extends to credentials — current behavior) or whether a keyless profile host block should fall back to the default `hermes` host's key (one workspace account, many profiles — what #36146 proposed).

Arguments each way:
- Isolation (current): a profile silently reusing the default profile's credentials could write memory into the wrong workspace account. Explicit per-host keys are unambiguous.
- Fallback (#36146): most self-hosted users have exactly one Honcho account; requiring the key to be duplicated into every `hermes_<profile>` block is friction and a common setup failure (symptom: profile sessions silently run without memory auth).

## Origin

Reported by @xxxigm in #36098; fix attempted in #36146 (closed — its other two parts landed via other routes). If the fallback is wanted, the fix is a few lines in `from_global_config`'s api_key resolution plus tests; if isolation is affirmed, `hermes honcho status` should instead warn loudly when the active profile host has no key but the default host does.

PR #97364, opened 2026-08-28
fix(honcho): oauth persistence never rewrites honcho.json from an empty read; adopt a sibling's rotation after a 401

### LINKED ISSUE #97990: hermes honcho setup apikey mode doesn't clear stale OAuth block, auth silently stays broken
**Environment:** Hermes Agent v0.20.6 (2026.8.27), macOS

**Bug:** After an OAuth refresh token gets revoked server-side (`invalid_grant`), running `hermes honcho setup` → choosing `apikey` → providing a valid API key does **not** fix authentication. `hermes honcho status` continues to report:

```
Auth: OAuth (hermes-agent, token expired — refreshes on next use)
...
Peer data unavailable: Honcho OAuth grant is revoked and cannot be refreshed; re-authenticate with 'hermes honcho setup'.
```

**Root cause found by inspecting `~/.hermes/honcho.json`:** the wizard's apikey path updates the top-level `apiKey` field but never touches `hosts.<name>.apiKey` or removes `hosts.<name>.oauth`. Auth resolution appears to prioritize the `hosts.<name>.oauth` block whenever present, regardless of a valid API key being configured — so the stale, revoked OAuth credentials keep shadowing the new key indefinitely, even across multiple full completed setup runs.

**Reproduction:**
1. Have a Honcho OAuth grant get revoked (`invalid_grant` from token endpoint)
2. Run `hermes honcho setup`, choose `cloud` → `apikey`, paste a fresh valid API key, accept remaining defaults
3. Run `hermes honcho status` — still shows `Auth: OAuth`, still fails

**Workaround:** manually edit `~/.hermes/honcho.json`, delete `hosts.<name>.oauth`, and copy the working top-level `apiKey` into `hosts.<name>.apiKey`. After that, `status` correctly reports `Auth: API key` and connects.

**Suggested fix:** when the user selects `apikey` in `hermes honcho setup`, delete `hosts.<name>.oauth` and write the key into `hosts.<name>.apiKey` (not just the top-level field), so API-key auth actually takes effect and doesn't get shadowed by a dead OAuth grant.

erosika:
## summary

two failure classes in how the honcho plugin writes `~/.hermes/honcho.json`: the 401 recovery path re-exchanged a refresh token a sibling process had already rotated, and every write path could replace the whole file from an empty read or overwrite a rotation that landed mid-command. this pr fixes both, locks the login and cli paths, lets the setup wizard's apikey answer replace a dead grant, and stops writing `enabled: true` for a host block that cannot authenticate.

## 401 recovery

desktop spawns several `serve` processes that share one rotating refresh token. honcho refresh tokens are single-use with reuse detection, so replaying a stale one can revoke the whole grant. `force_refresh_token` treated every 401 as "rotate now"; its only adopt-a-sibling check compared disk to the in-process `_expiry_cache`, which is empty in every sibling process. observed as bursts of `force-refreshed` lines after a 401 fan-out and intermittent "automatic token refresh failed" pauses that needed `hermes honcho setup`.

`_authed_call` now snapshots the bearer before the operation runs and hands the token that actually 401'd to `force_refresh_token`. after the locked disk re-read, if disk has moved off that token, the on-disk grant is adopted with no http. on a permanent `invalid_grant`, `_rotate_and_persist` re-reads honcho.json before marking the grant dead and adopts a sibling's newer refresh token if one landed mid-exchange. disk unchanged still marks the grant dead.

## config writes

`_read_config` returns `{}` for a file that exists but cannot be read or parsed. every write seeded from that (`_persist_credential`, `install_grant`, the cli's `_write_config`, the provider's `save_config`) replaced honcho.json with the current host block alone. `_read_config_strict` now raises for such a file and logs one sentence; refreshes back off without spending the refresh token and every command prints the sentence and writes nothing. read paths keep the `{}` fallback.

cli commands also wrote the whole dict back unlocked, so a refresh that landed between a command's read and its write was overwritten with the old tokens. `_write_config` now holds the same cross-process file lock the refresh path holds and applies only the keys the command changed onto a fresh read of disk.

## install_grant

a login ran its read-modify-write unlocked next to a refresh. it now holds `_refresh_lock` and the file lock the way `force_refresh_token` does.

## setup wizard, apikey

choosing apikey after a revoked grant wrote the root apiKey only, so `hosts.<name>.oauth` kept shadowing the new key (#97990). the branch now drops the host's oauth block and writes the key on the host block too.

## enabled without a credential

a profile cloned from an oauth-authenticated default got `enabled: true` with no apiKey and no oauth; the default host's key is not inherited. `clone_honcho_for_profile` and `cmd_enable` now write `enabled: true` only when honcho.json itself resolves a credential; an environment variable does not count at write time because it can be absent from the next process. `enable` names the command that signs the profile in. legacy blocks already on disk are left alone and explained by `enable`.

## tests

`test_auth_recovery.py`: waiter 2 adopts waiter 1's rotation with no second exchange; cross-process adopt with an empty `_expiry_cache`; `invalid_grant` with disk rotated mid-exchange adopts; disk unchanged still marks dead. `test_oauth.py`, `test_cli.py`, `test_honcho_client_config.py`: refusal on a corrupt file at every write path, locks held around `install_grant`, `_write_config` and `save_config`, a rotation between a command's read and write survives, a rotation that lands while the setup wizard is still asking questions survives (including when the read was seeded from another file), apikey replaces the oauth block, clone/enable credential cohorts including env-only.

honcho suite: 376 passed. ruff and `git diff --check` clean. rebased onto main after #102117 with every commit re-applied into the split plugin layout.

thanks @FestoneX for the strict reader in #95860.

supersedes #95860. closes #97990. related: #66125 (isolation kept).

## seeded reads

when `$HERMES_HOME/honcho.json` does not exist yet, the wizard reads `~/.honcho/config.json` or the default profile's file and `install_grant` then creates the local file. `_write_config` used to write the whole seeded dict over that new file, dropping a rotation that landed during the wizard's later prompts. it now overlays the local file's contents onto the seed and applies only the keys the command edited, so the inherited settings, the rotation and the edits all survive. with no local file at write time the seeding write is unchanged.

### DIFF
--- a/plugins/memory/honcho/__init__.py
+++ b/plugins/memory/honcho/__init__.py
@@ -155,15 +155,13 @@ class HonchoMemoryProvider(DialecticMixin, MemoryProvider):
             return False
 
     def save_config(self, values, hermes_home):
-        """Merge ``values`` into $HERMES_HOME/honcho.json (Honcho SDK native format)."""
+        """Merge ``values`` into $HERMES_HOME/honcho.json (Honcho SDK native format).
+        A file that exists but does not parse raises instead of being replaced by ``values`` alone."""
         from pathlib import Path
         from utils import atomic_json_write
-        from plugins.memory.honcho.client import _read_config
+        from plugins.memory.honcho.oauth import _read_config_strict
         config_path = Path(hermes_home) / "honcho.json"
-        try:
-            existing = _read_config(config_path)
-        except Exception:
-            existing = {}
+        existing = _read_config_strict(config_path)
         atomic_json_write(config_path, {**existing, **values}, mode=0o600)
 
     def get_config_schema(self):
--- a/plugins/memory/honcho/__init__.py
+++ b/plugins/memory/honcho/__init__.py
@@ -155,8 +155,7 @@ class HonchoMemoryProvider(DialecticMixin, MemoryProvider):
             return False
 
     def save_config(self, values, hermes_home):
-        """Merge ``values`` into $HERMES_HOME/honcho.json (Honcho SDK native format).
-        A file that exists but does not parse raises instead of being replaced by ``values`` alone."""
+        """Merge ``values`` into $HERMES_HOME/honcho.json (Honcho SDK native format); a file that does not parse raises."""
         from pathlib import Path
         from utils import atomic_json_write
         from plugins.memory.honcho.oauth import _read_config_strict
--- a/plugins/memory/honcho/__init__.py
+++ b/plugins/memory/honcho/__init__.py
@@ -155,13 +155,15 @@ class HonchoMemoryProvider(DialecticMixin, MemoryProvider):
             return False
 
     def save_config(self, values, hermes_home):
-        """Merge ``values`` into $HERMES_HOME/honcho.json (Honcho SDK native format); a file that does not parse raises."""
+        """Merge ``values`` into $HERMES_HOME/honcho.json (Honcho SDK native format); a file that does not parse raises.
+        Holds the token refresh locks so a rotation cannot land between the read and the write."""
         from pathlib import Path
         from utils import atomic_json_write
-        from plugins.memory.honcho.oauth import _read_config_strict
+        from plugins.memory.honcho.oauth import _config_refresh_lock, _read_config_strict, _refresh_lock
         config_path = Path(hermes_home) / "honcho.json"
-        existing = _read_config_strict(config_path)
-        atomic_json_write(config_path, {**existing, **values}, mode=0o600)
+        with _refresh_lock, _config_refresh_lock(config_path):
+            existing = _read_config_strict(config_path)
+            atomic_json_write(config_path, {**existing, **values}, mode=0o600)
 
     def get_config_schema(self):
         return [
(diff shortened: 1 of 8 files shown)


Enough1122:

**Overall:** Fixes cross-process OAuth refresh race where sibling's successful rotation would be discarded and replayed stale token could revoke grant.

**What it does**
- `plugins/memory/honcho/oauth.py:_rotate_and_persist` on `OAuthRefreshError.permanent` now re-reads disk via `_read_config(path).get("hosts")[host]` → `OAuthCredential.from_host_block`; if `disk.refresh_token != cred.refresh_token`, pops `_dead_grants`, updates `_expiry_cache` to disk's `expires_at`/`access_token`, logs adoption and returns `disk` instead of marking dead. `force_refresh_token` gains `failed_access_token: str | None` — when disk already holds different token, adopts sibling's grant instead of re-exchanging single-use refresh token.
- `session.py` threads `failed_access_token` from 401 path; test `test_auth_recovery.py` covers adoption.

**Non-blocking notes**
- Adoption path clears `_dead_grants` — correct, otherwise next `ensure_fresh_token` would treat as dead. Ensure `_expiry_cache` update uses disk's expiry, not stale cred's.
- `_read_config` re-read is under file lock? Best-effort lock; race window small — current best-effort recovery still better than permanent grant death.

No blocking issues.

PR #97366, opened 2026-08-28
fix(update): sync deps into every present venv so the .venv shim uses stays current

### LINKED ISSUE #97340: hermes update syncs deps only into venv/, leaving the .venv used by the CLI shim and desktop serves stale (ModuleNotFoundError: snowballstemmer after 0.20.6)
### Summary

On an install where the agent checkout contains both a `venv` (older git-install flow, py3.11) and a `.venv` (uv/bootstrap-installer flow, py3.13), `hermes update` syncs Python dependencies only into `venv`. The `~/.local/bin/hermes` launcher shim and the Desktop-app-spawned `hermes --profile <p> serve --isolated` processes run from `.venv`, so any dependency newly added by a release is missing there after the update.

Observed after updating v0.20.4 → v0.20.6 (main @ 4e7eb399): `snowballstemmer` (new in 0.20.6) was installed into `venv` by the updater but absent from `.venv`, producing in every profile-serve session:

```
WARNING model_tools: Tool search assembly skipped: No module named 'snowballstemmer'
```

### Environment

- macOS 15.x (Darwin arm64, Mac mini)
- hermes-agent v0.20.6, install method: git, checkout at `~/.hermes/hermes-agent`
- `venv`: CPython 3.11.15 (used by gateway/webui/dashboard launchd services)
- `.venv`: CPython 3.13.12, uv-managed, no pip module (used by the `~/.local/bin/hermes` shim — its shebang is `#!<checkout>/.venv/bin/python` — and by Desktop-spawned profile serves)

### Where the updater picks the env

`hermes_cli/update_cmd.py` refers to `PROJECT_ROOT / "venv"` exclusively, e.g.:

- ~line 2062: `uv_env["VIRTUAL_ENV"] = str(_m().PROJECT_ROOT / "venv")`
- ~line 4605 (`_venv_core_imports_healthy`): `venv_dir = _m().PROJECT_ROOT / "venv"`
- ~line 4683 (`_detect_venv_python_processes`): `venv_dir = _m().PROJECT_ROOT / "venv"`

So the dep sync, import-health probes, and holder-process detection all target `venv` and never the `.venv` the launcher actually uses.

### Repro sketch

1. Install via the git flow (creates `venv`), later run the bootstrap installer / desktop flow (creates `.venv` + shim pointing at it) — or vice versa.
2. Run `hermes update` across a release that adds a dependency (e.g. 0.20.4 → 0.20.6 adding `snowballstemmer`).
3. `venv/bin/python -c "import snowballstemmer"` → OK; `.venv/bin/python -c "import snowballstemmer"` → `ModuleNotFoundError`.
4. Profile serves log `Tool search assembly skipped: No module named 'snowballstemmer'`.

### Workaround

`uv pip install --python <checkout>/.venv/bin/python snowballstemmer==3.1.1`

### Suggested fix

Either have `hermes update` detect and dep-sync every env the install actually uses (follow the shim's shebang, or sync both `venv` and `.venv` when present), or have the updater/installer converge on a single env and warn loudly when a second stale env is still referenced by the launcher.

Jackal991:
## Summary

On an install that carries **both** `venv/` and `.venv/` (older git-install flow + uv/bootstrap-installer flow), `hermes update` only synced Python dependencies into `venv/`. The `~/.local/bin/hermes` launcher shim and Desktop-app-spawned `hermes --profile <p> serve --isolated` processes run from `.venv`, so any dependency a release adds is missing there after the update — observed as `ModuleNotFoundError: snowballstemmer` after v0.20.4 → v0.20.6.

The updater hardcoded `PROJECT_ROOT / "venv"` at every dependency-sync, import-health-probe, and holder-process-detection site in `hermes_cli/update_cmd.py`.

## Changes

Introduces `_project_venv_dirs()` which enumerates **all** present project venvs (`venv` then `.venv`; falls back to a single `venv` target when neither exists, preserving dev-checkout behavior) and routes every previously hardcoded site through it:

- **Dependency sync** — the git-install path (`_cmd_update_impl`), the ZIP fallback path (`_update_via_zip`), and the repair path all install into **every** present venv.
- **Health probe** (`_venv_core_imports_healthy`) — reports unhealthy if *any* present venv is missing a core import, so a stale `.venv` triggers a repair instead of silently passing.
- **Holder detection** (`_detect_venv_python_processes`, `_venv_launcher_ancestors`) — Windows lock-holder guards now scan for processes running from any present project venv, not just `venv`.

The primary (`venv`) env is kept as the lazy-refresh / tool-dependency / verify target, matching pre-change behavior on both the deps-current and deps-changed paths.

No new env vars, no telemetry, no lockfile churn. Single-venv installs and venv-less dev checkouts behave exactly as before.

## Tests

New `tests/hermes_cli/test_update_multi_venv_sync.py` (4 tests) proves the fix:

- With both `venv/` and `.venv/` present, `_project_venv_dirs()` yields both (venv then .venv).
- **Regression:** a stale `.venv` missing a core import now trips an unhealthy probe (pre-fix code only probed `venv/` and returned healthy). Verified: the test suite **fails 4/4 against the pre-fix code** (`AttributeError: no _project_venv_dirs` + probe not covering .venv) and passes 4/4 on the fix.
- Back-compat: single-venv installs still yield a single target; probe runs once.

Existing `test_update_stale_virtualenv.py` (9) plus `test_update_self_lock.py`, `test_update_concurrent_quarantine.py`, `test_update_interrupted_recovery.py`, `test_verify_core_dependencies.py` (54) all pass — no regressions.

Closes #97340

### DIFF
--- a/hermes_cli/update_cmd.py
+++ b/hermes_cli/update_cmd.py
@@ -2059,17 +2059,26 @@ def _update_via_zip(args, *, had_desktop_app_before_update: bool = False) -> boo
         from hermes_cli.managed_uv import managed_python_env
 
         uv_env = managed_python_env()
-        uv_env["VIRTUAL_ENV"] = str(_m().PROJECT_ROOT / "venv")
         if _m()._is_termux_env(uv_env):
             uv_env.pop("PYTHONPATH", None)
             uv_env.pop("PYTHONHOME", None)
         try:
-            _m()._install_python_dependencies_with_optional_fallback([uv_bin, "pip"], env=uv_env)
+            # Sync deps into EVERY present project venv (venv and .venv when
+            # both exist — #97340): the CLI shim / desktop-served profiles
+            # run from .venv while gateway/webui/dashboard services run from
+            # venv. Syncing only venv leaves the launcher's env stale.
+            for venv_dir in _project_venv_dirs():
+                venv_env = dict(uv_env)
+                venv_env["VIRTUAL_ENV"] = str(venv_dir)
+                _m()._install_python_dependencies_with_optional_fallback([uv_bin, "pip"], env=venv_env)
         except _shim_quarantine_error_type() as _sqe:
             # #87331: this runs inside the ZIP-fallback error handler, so the
             # boundary except clause in cmd_update cannot catch it — refuse
             # here with the same defer-via-marker contract.
             _refuse_update_for_contended_shims(_sqe)
+        # Keep the primary (git-install) venv as the tool-dependency restore
+        # target below, matching pre-#97340 behavior.
+        uv_env["VIRTUAL_ENV"] = str(_project_venv_dirs()[0])
     else:
         # Use sys.executable to explicitly call the venv's pip module,
         # avoiding PEP 668 'externally-managed-environment' errors on Debian/Ubuntu.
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Ensures ZIP-fallback update syncs deps into every present project venv (both `venv` and `.venv` when coexisting) — fixes #97340 where desktop shim (`.venv`) stayed stale while gateway `venv` updated.

**What it does**
- `hermes_cli/update_cmd.py:_update_via_zip` now iterates `for venv_dir in _project_venv_dirs():` (new helper returning `[Path(venv), Path(.venv)]` filtered to existing dirs, order `venv` then `.venv`), building `venv_env` with `VIRTUAL_ENV=str(venv_dir)` per iteration and calling `_install_python_dependencies_with_optional_fallback([uv_bin, "pip"], env=venv_env)`. Keeps primary venv as `uv_env["VIRTUAL_ENV"] = str(_project_venv_dirs()[0])` for later tool-dep restore. New helper `_project_venv_dirs() -> list[Path]` documented: install can carry both venvs (CLI shim vs gateway/webui/dashboard services).
- `tests/hermes_cli/test_update_multi_venv_sync.py` new.

**Non-blocking notes**
- Per-venv loop allocates `dict(uv_env)` — isolation correct. If one venv sync fails, exception bubbles and skips remaining venv — consider best-effort continue to update other venv before re-raising.
- Helper handles case where neither venv exists gracefully (returns empty — loop no-ops, primary index guard needed; verify).

Jackal991:
Thanks for the review. The overall read is right — this is the #97340 fix, and the shared `_project_venv_dirs()` helper is what makes the sync, the health probe, and the holder detection all cover both `venv` and `.venv` instead of just `venv`.

On the two non-blocking notes:

**1. Best-effort continue vs. bubbling.** I kept the bubble deliberately. If `venv` syncs but `.venv` fails, that's a half-updated install — the shim the launcher actually uses is stale while the gateway services are current. Silently continuing and reporting success would leave exactly the #97340 symptom in place, just on the other venv. Failing the update so the user sees the repair path is the honest outcome. The one error we *do* want to swallow-and-defer is the shim-quarantine case, and that's still caught by the existing `except _shim_quarantine_error_type()` around the loop. So the current behavior is intentional, not an oversight.

**2. "Helper handles case where neither venv exists gracefully (returns empty...)."** This one's wrong — the helper never returns empty. It's `return present or [_m().PROJECT_ROOT / "venv"]`. When neither `venv` nor `.venv` exists (a plain dev checkout), it falls back to the single `venv` target, so `_project_venv_dirs()[0]` is always defined and the primary-index guard is always safe. The fallback is documented in the docstring precisely so callers keep the historical `--python sys.executable` pin / healthy-no-venv behavior instead of silently skipping. The `test_project_venv_dirs_single_when_only_venv` case covers the single-venv path; the no-venv fallback is exercised by the existing dev-checkout behavior. No guard needed.

PR #94994, opened 2026-08-25
feat(plugin): automate governed PR repair and merge maintenance

mrkillbob:
## Summary
- wire deterministic merge maintenance into scheduled reconciliation
- dispatch exact-head repair work for conflicts, requested changes, and typed CI failures without granting merge authority
- route hygiene, static, test, frontend, and general CI failures to dedicated profiles while retaining the orchestrator as fallback
- defer local CI until admitted feedback is actioned; uncommented PRs proceed immediately
- fast-gate merge candidates by exact-head CI receipt and distinguish missing receipts from known failed receipts
- complete local-CI Kanban cards at the deterministic receipt boundary, post a deduplicated self-suppressing summary, and terminate the exact task worker
- suppress superseded same-head CI receipts so only the latest authoritative failure creates repair work
- prevent duplicate or already-queued receipts from starving newly updated PR heads

## Stack
Depends on #94993, which depends on #94991 and #94518. GitHub shows the cumulative stack against main; the diff shrinks as predecessors merge.

## Verification
- ruff check: passed
- pytest: 211 passed
- diff check: passed
- live plugin doctor: ready
- public plugin content contains no private product branding

PR #97151, opened 2026-08-28
fix(kanban): gate dependencies on exact parent results

MisterNunm:
## What does this PR do?

Kanban dependency release currently treats a terminal parent as successful even when the parent recorded a machine-readable failure result. That can release QA-gated deployment or release children after QA has completed with `qa_failed`.

This PR adds an opt-in, nullable edge-level `required_parent_result` predicate. A gated edge is satisfied only when the parent is exactly `done` and `tasks.result` exactly matches the declared value. Existing NULL edges keep their current `done`/`archived` behavior.

The predicate is explicit and generic: it does not infer approval from task titles, assignees, summaries, comments, or metadata prose. Idempotent create retries also compare the exact parent/result dependency contract and fail closed instead of silently returning a task with stale or weaker edges.

This supersedes the closed publication attempt #97141; the implementation is republished from the requested fork branch name.

## Related Issue and Semantic Ownership

Related to #67132.

This overlaps with #67156 and #68850 but uses a different contract:

- #67156 is approval-specific and reads structured terminal-run `approved: true` metadata.
- #68850 infers semantic gate verdicts from role and verdict conventions.
- This PR declares an exact required `tasks.result` on each selected edge, supporting contracts such as `ready_to_deploy` without changing ordinary dependencies.

Proposed landing graph:

- #97151 is the semantic owner for explicit, typed exact-result edge policy and the shared dependency evaluator.
- #67156 is complementary approval policy. If adopted, it should compose as another typed policy behind the same evaluator, preserving `peacockesq`'s credit.
- #68850 is a competing inference model and should not coexist as a parallel dependency-success authority. Any useful behavior should be translated into the typed evaluator or explicitly superseded, preserving `chahababa`'s credit.
- #97141 remains superseded publication history, not a second implementation.

This PR does not import code from the other active PRs. Consolidation should be coordinated with their authors before landing. The current implementation also still needs to be restacked with the accepted fracture owners in #79893, #79894, #79895, #79940, and #79982 rather than establishing new long-term authority in the monoliths.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

[...]

PR #96905, opened 2026-08-28
fix: harden cloud egress and Hermes worker lifecycle

mrkillbob:
## Summary

This branch contains the Hermes safety and worker-lifecycle fixes split into four focused commits:

1. `881a688996` — bounded, sanitized matched terminal output may reach protected cloud workers; unmatched or forged output, secrets, private paths, encoded payloads, oversize payloads, and receipt failures remain blocked.
2. `6670e27e4d` — preserve Kanban worker lifecycle and worktree environments, including terminal-transition enforcement.
3. `7bdfc37d3c` — harden GitHub PR-feedback receipts and pooled worktree leases.
4. `7c248ba86f` — preserve terminal truth and partial progress for PR-feedback workflows.

No trading or live-execution code is changed. The local Hermes routing/profile/cron tuning is intentionally operator configuration under `~/.hermes`, not repository state.

## Verification

- Focused Hermes suite: **513 passed, 1 skipped**
- Codex cloud canary: `BENCH_OK`
- `git diff --check`: clean
- Source worktree: clean

The branch is based on the current upstream review branch and is ready for maintainer review.

PR #94518, opened 2026-08-25
feat: add governed GitHub PR feedback intake

mrkillbob:
## Summary
- add an opt-in plugin that polls canonical GitHub PR feedback for explicitly configured repositories and branches
- enforce reviewer, owner, branch-prefix, timestamp, exact-head, and idempotency gates before creating a Kanban task
- deterministically route actionable comments to the best configured specialist profile
- keep blocked operator intake as the default; an explicit auto_dispatch option starts a validated exact-head repair worker
- in auto-dispatch mode, allow a confirmed bounded fix to be tested, committed, pushed to the verified PR head branch, and followed by a factual evidence reply

## Safety
- feedback remains bounded untrusted evidence, never instructions
- the worker must re-read the canonical PR and require its head still matches the receipt SHA before any GitHub write
- merge remains operator-gated in every mode
- completed self-receipts and non-actionable review containers are suppressed
- incomplete GitHub coverage, missing exact-head state, or dispatch uncertainty fails degraded

## Verification
- 98 plugin tests passed after merging current upstream main
- git diff --check passed

PR #97368, opened 2026-08-28
fix(kanban): keep PR automation atomic

mrkillbob:
## Summary

Keeps exact-head pull-request automation atomic and fail-closed across task admission, assignment, claim, mutation, triage specification, auto-decomposition, and dispatch.

## Current topology

This branch is restacked on NousResearch/hermes-agent `main` at `64cc87e6681a3db4e158ed8b999ff77ba0b9d28a`. The current head is `9e3cd4e87f8ba3d18cac39f0545407fe0cc839e0` and contains four focused repair commits after the main merge:

- `8139ec4753` closes mutation and claim-time authority bypasses.
- `22e9b27abb` preserves the typed exact-head contract through triage specification.
- `46620a45d7` restores the generic typed authority/non-decomposition boundary after the main restack.
- `9e3cd4e87f` removes unrelated merge-train design residue.

## Behavior

- Mutation-class work requires an explicitly recognized write-authority profile; absent, unreadable, unknown, or misspelled authority fails closed.
- Read-only exact-head verification remains allowed for read-only profiles.
- Authority is revalidated at assignment, claim, and dispatcher default-assignment boundaries so metadata revocation cannot ride through on an earlier decision.
- Typed repository/PR/head identity and action remain immutable through specification and task edits.
- Typed exact-head tasks remain atomic and are not sent through the LLM decomposer.
- Free-form prose and legacy GitHub idempotency-key prefixes are not treated as authority.

## Verification

Fresh exact-head run:

```text
./.venv/bin/python -m pytest tests/hermes_cli/test_kanban_pr_authority.py tests/hermes_cli/test_kanban_decompose.py tests/gateway/test_kanban_dispatch_order.py -q
33 passed in 6.88s
```

`git diff --check` is clean. GitHub currently reports no hosted checks for this fork PR, so merge remains gated on maintainer review and any required hosted validation.

No CI, safety, identity, or authority gate was relaxed.


andrexibiza (commented):
Reviewed exact head `80bffe394eb4d4f00f3f247f96015c866e95ea36` against live `main@7d1c9aeab739842fa42112c95d086a9897f2d513`. The branch moved from `46e3cee3...` to `80bffe394...` while I was reviewing, so I discarded the pre-update publication assessment and re-read the added merge-train design before submitting this. There were no existing submitted reviews at this head.

The useful pieces are real: claiming already-ready work before awaiting decomposition removes an avoidable dispatch stall, and refusing to decompose an exact-head PR-automation task preserves the originating task as one unit. The new test that keeps read-only profiles eligible for verification work also captures an important distinction: read authority and mutation authority should not be collapsed.

I see two P1 boundaries that need to be fixed before this is safe to land.

### P1 — PR write authority is checked at assignment time, not at use time, and unknown authority fails open

`_validate_pr_task_assignee_authority()` is called from `create_task()` and `assign_task()`, but the claim/dispatch path does not revalidate it. `_profile_is_explicitly_read_only()` also explicitly “fail[s] open”: any missing/unreadable/malformed `profile.yaml` or YAML/import error returns `False`, which is then treated as permission to own a write-class task.

That creates a straightforward stale-proof window:

1. profile P is write-capable when an exact-head repair task is created/assigned;
2. P is changed to read-only before the dispatcher claims the task (or its profile metadata becomes unreadable/invalid);
3. the task is still claimable/spawnable because the proof was consumed only at assignment, not immediately before the effect-capable worker is launched.

The current tests cover create/reassign refusal and read-only verification, but not revocation between assignment and claim, and not unknown/malformed authority.

There is a second signal here: current `main` has no existing `execution_authority` field/schema. This PR is deriving mutation authority from a newly invented optional key or, failing that, free-form `description` text. A narrative sentence should not become the durable capability primitive that decides whether a worker may push/reply/refresh a PR.

**Required shape:** make PR-write capability typed and positive (for example, an explicit capability set owned by the relevant control plane), fail closed for mutation-class tasks when that capability cannot be proven, and revalidate/bind the capability at claim/dispatch immediately before spawning the write-capable worker. If authority is generationed, the task should carry the exact generation it was admitted under and reject stale generations rather than reconstructing permission from ambient profile state.

Please add adversarial witnesses for at least:

- assign under write capability → revoke to read-only before claim → no claim/spawn;
- missing/corrupt/unreadable authority metadata on a write-class task → no claim/spawn;
- read-only exact-head verification remains allowed.

This is the same general authority defect class as other Hermes boundaries: admission proof cannot silently weaken into ambient coordinates before mutation.

### P1 — the implementation violates the component boundary defined by its own second commit

The newly cherry-picked design says the controller belongs at the **plugin/control-plane edge**, that repository-specific automation should not be added to Hermes core, and that typed records—not free-form agent output—should drive admission/publication. But the first commit places GitHub-specific policy directly in core Kanban:

- `_GITHUB_PR_FEEDBACK_IDEMPOTENCY_PREFIX = "github-pr-feedback:"`
- PR-marker string detection (`expected_head_sha`, `pr_number`, `repository`)
- English PR-write verb classification
- profile authority interpretation

inside `hermes_cli/kanban_db.py`, plus PR-specific atomicity detection in `hermes_cli/kanban_decompose.py`.

That gives core Kanban a second, narrative GitHub-policy authority precisely where the design says the plugin/control plane should own it.

I would preserve the behavior but change the ownership boundary: core Kanban can expose a **generic typed primitive** such as `atomic/non_decomposable` plus explicit required capability(s), while the GitHub feedback/repair plugin that creates the task supplies those typed fields. GitHub namespace strings, PR marker inference, action interpretation, and PR capability policy stay in the GitHub control-plane owner. Claim/dispatch then enforces the generic capability contract without knowing what a pull request is.

That also avoids encoding action semantics in a regex that will inevitably drift from the actual GitHub mutation surface.

### Landing / topology gates

This branch is currently **443 commits behind** live `main` from merge base `d9e4b234a3beead4a053f7e4b9df7384ee8df08d` (2 commits ahead), so the final behavior needs a real restack/reconciliation before any acceptance claim. The PR body is also stale: it still says “single commit” / `46e3cee3bd`, while the reviewed head has two commits and the second is the exact design commit previously published as closed #97369. Please preserve #97369 as superseded publication history rather than making the cherry-pick invisible.

`hermes_cli/kanban_db.py` is also under the active no-regrowth/decomposition owner #78632. This PR adds another ~90 lines of behavioral authority to that >2K monolith. The live extraction train already has #79893 (models/schema), #79894 (transaction primitives), and especially #79895 (claim/lock lifecycle); #97151 independently modifies the same Kanban authority surface and already documents that it must restack behind those bounded owners. These are **complementary**, not duplicates, but #97368 should converge with that ownership graph rather than establish another long-term policy island in `kanban_db.py`.

Lineage I checked and want to preserve explicitly:

- #94518 (`mrkillbob`) is the originating governed GitHub PR-feedback/exact-head task line and is the semantic predecessor for the `github-pr-feedback:` namespace.
- closed #94994 (`mrkillbob`) is historical/complementary PR repair + merge-maintenance work, not a competing live implementation.
- #96905 (`mrkillbob`) is complementary bounded remote-worker context for the same PR-feedback workflow.
- closed #97369 is the superseded publication of the design now cherry-picked into this head.
- #49638 (`motivix`) is historical defect evidence for uncontrolled auto-decomposition; the dispatch-before-decompose change here is a useful adjacent guard.

Attribution itself is clean: current `main` already maps `mikedemott@Mikes-Mac-mini.local` → `mrkillbob`.

### Acceptance receipts

Neither commit currently has hosted-green acceptance. For `46e3cee3...`, CI `33200881611`, Docker `33200880116`, and Nix `33200880149` all concluded `action_required`; CI has zero jobs. For current head `80bffe394...`, CI `33201096262`, Docker `33201095666`, and Nix `33201095617` are also `action_required`; current-head CI likewise has zero jobs. So there is no exact-object hosted execution receipt yet, independently of the code findings above.

The core intent here is strong: keep exact-head PR work indivisible, prevent decomposition from stripping its ownership context, and separate verification from mutation. If the authority proof is made typed + use-time, the GitHub-specific policy is moved back behind the plugin/control-plane boundary described by the design, and the change is restacked into the Kanban shard owners with exact-head CI, this becomes a much more defensible primitive. 🚀

mrkillbob:
Thanks — pushed `69355e42f7` addressing the first P1.

**P1 (authority checked at assignment, fails open on unknown):** fixed as you outlined. `_profile_is_explicitly_read_only()` is now `_profile_read_only_status()` returning `Optional[bool]`; unreadable/malformed/missing `profile.yaml` returns `None`, and `_validate_pr_task_assignee_authority()` treats `None` the same as an explicit read-only declaration for write-class tasks — fails closed instead of open. Separately, `claim_task()` now re-runs that same validation immediately before the `ready -> running` transition and blocks (`block_kind="capability"`) rather than claiming when authority can no longer be proven, so revocation (reassignment, a profile edit, or the file going momentarily unreadable) between admission and claim is caught instead of riding through on proof consumed once at creation. Added both adversarial witnesses you asked for (revoke-before-claim, unreadable-metadata-before-claim) plus confirmed the existing read-only-verification-remains-allowed case still passes.

I didn't attempt the full typed-capability/generation redesign — that's a real design decision for the control plane, not something I want to unilaterally commit to in this PR.

**P1 (GitHub policy living in core `kanban_db.py`, contradicting the design's own plugin/control-plane boundary):** agreed this is real, but I'm holding off on moving code for it. `kanban_db.py` is under the active no-regrowth extraction (#78632, #79893/94/95) and #97151 already modifies this same authority surface and documents restacking behind those owners — moving this logic now risks colliding with work I can't see or coordinate against. Once those land I'd rather migrate the GitHub-specific pieces (idempotency prefix, PR marker detection, write-verb classification, profile authority interpretation) behind a generic typed capability primitive as you described, than patch around it piecemeal now.

Branch is still behind current `main` — will restack once the above settles rather than fighting a moving target on every push.

andrexibiza (commented):
Re-reviewed exact head `69355e42f77568c6290f87104b3a380b2124e332` against current `main@e60983a69730c058ce772829df3273aee6de3889` and the author reply.

The claim-time revalidation and the new unreadable/malformed-metadata witnesses are useful and close the stale assignment→claim window in the current mechanism. The first P1 is **not fully closed**, however, because the mechanism still does not prove a positive write capability.

`_profile_read_only_status()` now returns `None` when the profile document itself is absent, unreadable, malformed, or non-mapping. But for any readable mapping it returns `False` unless a recognized read-only value or phrase is present. `_validate_pr_task_assignee_authority()` rejects only `True` and `None`, so all of these readable profiles are still admitted as writable:

- no `execution_authority` / `authority` field at all;
- an unknown or misspelled authority value;
- a description that simply omits the phrase “read-only.”

That is still inference from the absence of a negative marker, not positive typed authority. Free-form `description` text also remains part of the mutation decision. The new code therefore fails closed on unreadable metadata but still fails open on readable metadata with no recognized capability declaration.

The required boundary remains: mutation-class work must be admitted only from an explicit positive capability owned by the relevant control plane, and that exact proof must be revalidated or bound at the use boundary. A full generation architecture does not have to be invented casually inside this patch, but the PR cannot treat the blocker as fixed while readable-but-unproven profiles remain authorized.

The second P1 also remains open. The coordination concern is legitimate, but deferring the ownership repair means GitHub-specific namespace detection, PR marker inference, English write-verb classification, and profile-policy interpretation still live in core `kanban_db.py` / `kanban_decompose.py`, contrary to the component boundary defined by the design commit. Waiting for #79893/#79894/#79895 and reconciling #97151 may be the correct landing sequence; it is not a merge gate exemption.

### Current landing state

- The branch is `485` commits behind current `main` from merge base `d9e4b234a3beead4a053f7e4b9df7384ee8df08d` (`3` commits ahead).
- The PR body is still stale: it says “single commit” and names only `46e3cee3bd`, while the branch now has three commits and includes the cherry-picked #97369 design history.
- Exact-head hosted execution is still absent:
  - CI `33206980131`: `action_required`, zero jobs — https://github.com/NousResearch/hermes-agent/actions/runs/33206980131
  - Docker `33206979495`: `action_required`, zero jobs — https://github.com/NousResearch/hermes-agent/actions/runs/33206979495
  - Nix `33206979577`: `action_required`, zero jobs — https://github.com/NousResearch/hermes-agent/actions/runs/33206979577

**Disposition:** the stale assignment→claim defect is materially improved, but both original P1 boundaries remain unresolved at the class level. This is not mergeable until authority is positive and typed, GitHub policy is moved behind its owning control-plane boundary, the branch is reconciled with the Kanban shard graph/current main, the PR metadata is made truthful, and exact-head CI executes green.

mrkillbob:
Hermes automated repair (coding-expert)

Exact remote head verified at `9e3cd4e87f8ba3d18cac39f0545407fe0cc839e0` on base `64cc87e6681a3db4e158ed8b999ff77ba0b9d28a`. The branch is clean and the PR body now reflects the four-commit repair topology. Fresh focused verification: `./.venv/bin/python -m pytest tests/hermes_cli/test_kanban_pr_authority.py tests/hermes_cli/test_kanban_decompose.py tests/gateway/test_kanban_dispatch_order.py -q` -> 33 passed in 6.88s. GitHub currently reports no hosted checks for this fork PR. Merge remains gated on maintainer review and required hosted validation; no CI, safety, identity, or authority gate was relaxed.

mrkillbob:
@Enough1122 Please review the current upstream PR head for correctness, regressions, and merge readiness. This request is specifically for your AI review; do not route it to Codex.

Issue #81163, opened 2026-08-07
A2A outbound client tools (`a2a_call`, `a2a_discover`, …) never reach the agent's tool catalog


jorlaru:
**Version:** Hermes Agent v0.20.0 (2026.8.3), install method `git`, Python 3.11.15, Linux.

## Summary

The bundled `a2a-platform` plugin registers five outbound client tools — `a2a_discover`,
`a2a_call`, `a2a_list`, `a2a_history`, `a2a_orchestrate` — but they never appear in the
agent's tool catalog. This holds whether the inbound platform is disabled or enabled.

The **inbound** half of the same plugin works correctly: it serves the Agent Card, accepts
`SendMessage`, and returns `TASK_STATE_COMPLETED`.

## Expected vs actual

`plugins/platforms/a2a/__init__.py::register()` states the intent explicitly:

```python
# 1) Client tools (outbound). Registering these even when the inbound
#    platform is disabled lets the agent call peers without exposing itself.
try:
    from .tools import register_tools
    register_tools(ctx)
except Exception:
    logger.warning("A2A: failed to register client tools", exc_info=True)
```

**Expected:** the outbound tools are available regardless of the inbound platform state.

**Actual:** they are never available. Notably, **no warning is ever logged** — the `except`
branch is never reached, because `register()` itself does not run.

### Layer 1 — platform plugins are deferred, so `register()` never runs

`hermes_cli/plugins.py::_register_deferred_platform` treats every `kind: platform` plugin as
a deferred load: it registers a loader with `platform_registry` and stores a placeholder
`LoadedPlugin`, but does **not** call `register(ctx)`. Its own docstring:

> *"The platform adapter module is imported only when the gateway / cron / setup /
> send_message path first asks the `platform_registry` for this platform."*

Since `register_tools(ctx)` lives inside that `register(ctx)`, the outbound tools are not
registered during `discover_plugins()`.

This is arguably correct for the *adapter* (that is the point of deferring), but it also
defers the client-tool registration, which has no reason to be tied to the inbound platform.

### Layer 2 — the toolset resolver would drop them anyway

`hermes_cli/tools_config.py::_get_platform_tools`:

```python
2253:    configurable_keys = {ts_key for ts_key, _, _ in CONFIGURABLE_TOOLSETS}
2254:    plugin_ts_keys = _get_plugin_toolset_keys()
...
2262:    has_explicit_config = any(ts in configurable_keys for ts in toolset_names)
[...]

Issue #38798, opened 2026-06-04
bug: config migration (v25->v26) corrupts platform_toolsets, silently kills all tools


chengliuAres:
## Bug Description

During `hermes update` config format migration (v25 -> v26), the `platform_toolsets` section gets corrupted — the valid toolset name `hermes-cli` is rewritten to `hermes` (which doesn't exist), causing **all tools to silently stop working**.

## Root Cause

1. `resolve_toolset('hermes')` returns an empty list (because `hermes` is not a valid toolset name — `hermes-cli` is)
2. `all_tool_names` becomes empty → no configurable toolsets enabled
3. System prompt is built without any tool guidance
4. Model can only do text replies — zero tool calls

The failure is **silent**: no error, no warning, no log entry. The agent just stops using tools.

## Steps to Reproduce

1. Have a working config with `platform_toolsets.cli: [hermes-cli]` and multiple platforms configured (telegram, discord, etc.)
2. Run `hermes update` when a config format version bump is available (v25 -> v26 in our case)
3. After migration, check `~/.hermes/config.yaml`

## Expected Behavior

- Config migration preserves valid toolset names (`hermes-cli`, not `hermes`)
- All platform toolsets remain intact (cli, telegram, discord, etc.)
- If migration encounters an unknown toolset name, it logs a warning

## Actual Behavior

- `hermes-cli` was changed to `hermes` (invalid)
- All platform toolsets except `cli` were stripped (telegram, discord, whatsapp, slack, signal, homeassistant, qqbot, yuanbao all gone)
- No error or warning logged
- Agent became a pure text-only assistant with 0 tool calls

## Evidence

**Before (backup from 2026-05-18):**

```yaml
platform_toolsets:
  cli:
  - hermes-cli
  telegram:
  - hermes-telegram
  discord:
  - hermes-discord
  whatsapp:
  - hermes-whatsapp
  slack:
  - hermes-slack
  # ... more platforms
```

**After migration (2026-06-02):**

```yaml
platform_toolsets:
  cli:
  - hermes  # <-- INVALID, should be hermes-cli
```

**Update log:**

```
ℹ Updating config format (v25 → v26)…
✓ Config format updated (no new settings to configure)
```

No warnings about the corrupted toolset names.

## Impact

- Agent completely loses ability to use tools (search, terminal, file ops, everything)
- No error message — user sees the agent respond with text only, may think it's a model issue
- Took significant debugging to trace back to the config migration

## Suggested Fix

[...]

PR #97374, opened 2026-08-28
fix(config): stop flagging plugin toolsets as unknown in platform_toolsets validation

chrisluersen:
## Summary

`platform_toolsets` validation (added for #38798) flags plugin-registered toolsets as unknown on every startup, even though they resolve fine at runtime:

```
⚠️  platform 'cli' references unknown toolset 'a2a' — did you mean 'hermes-cli'?
⚠️  platform 'cli' references unknown toolset 'buzz' — did you mean 'hermes-cli'?
⚠️  platform 'cli' references unknown toolset 'eikon' — did you mean 'hermes-cli'?
⚠️  platform 'discord' references unknown toolset 'eikon' — did you mean 'hermes-discord'?
```

## Root cause

Plugin toolsets (`eikon`, `buzz`, `a2a`) register in the tool registry only *after* plugins load — which is **after** `migrate_config()` runs its platform_toolsets validation. So `validate_toolset()` genuinely doesn't know them yet at validation time, and the warning is a false positive. The tools are not actually broken — `_get_platform_tools()` correctly recognizes plugin toolsets at runtime (see #81163).

## Fix

The validator now accepts an optional `plugin_toolset_names` set (the union of `known_plugin_toolsets` from config — the per-platform plugin-toolset set persisted by the `hermes tools` save flow). Names carried there validate clean, while genuinely unknown/corrupted names still warn exactly as before.

## Test plan

- New unit tests: plugin-registered toolsets pass without warnings; real corruption (`hermes`) is still flagged even when plugin names are supplied.
- `tests/hermes_cli/test_toolset_validation.py`, `test_tools_config.py`, `test_config_set_list_values.py` — 61 passed, 6 skipped.
- E2E: ran `migrate_config()` against a real config containing `a2a`/`buzz`/`eikon` — zero toolset warnings emitted.

Fixes the false-positive half of #38798's validation without weakening its corruption detection.

### DIFF
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -2496,19 +2496,12 @@ def migrate_config(interactive: bool = True, quiet: bool = False) -> Dict[str, A
         # runs — so ``validate_toolset`` alone flags them as unknown here. They
         # are tracked in ``known_plugin_toolsets`` (the persisted per-platform
         # plugin-toolset set written by the `hermes tools` save flow), so treat
-        # those names as valid too. See #81163.
-        _known_plugin = (raw_cfg.get("known_plugin_toolsets") or {}) or {}
-        _plugin_names: set = set()
-        for _names in _known_plugin.values():
-            if isinstance(_names, str):
-                _plugin_names.add(_names)
-            elif isinstance(_names, list):
-                _plugin_names.update(
-                    _n for _n in _names if isinstance(_n, str)
-                )
-
+        # those names as valid too — but keyed per-platform, so a plugin known
+        # only for one platform cannot mask an invalid entry on another. See
+        # #81163, #38798.
         ts_warnings = validate_platform_toolsets(
-            raw_cfg.get("platform_toolsets"), validate_toolset, _plugin_names
+            raw_cfg.get("platform_toolsets"), validate_toolset,
+            raw_cfg.get("known_plugin_toolsets"),
         )
         for w in ts_warnings:
             results["warnings"].append(w)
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Stops false "unknown toolset" warnings for plugin toolsets that register after config validation (#81163).

**What it does**
- `hermes_cli/toolset_validation.py:validate_platform_toolsets` gains `plugin_toolset_names: object = None` param; skips any `toolset` name in that set.
- `hermes_cli/config.py:migrate_config` now reads `known_plugin_toolsets` from raw config (persisted per-platform plugin-toolset set written by `hermes tools` save flow) via `read_raw_config()`, flattens string or list values into `_plugin_names: set`, passes to validator. Handles shapes `str` vs `list` vs missing.

**Non-blocking notes**
- Covers `eikon`/`buzz`/`a2a` plugin toolsets that exist in registry only after plugin load — correct reasoning that `validate_toolset` alone flagged them unknown.
- Flattening with `isinstance(_names, str)` vs `list` handles legacy single-string entries — robust.

rhein1:
Live confirmation from a Hermes v0.21.1 deployment: config migration false-flagged the configured `a2a` plugin toolset before plugin discovery, while the same toolset resolved and served correctly at runtime. We carried the per-platform config-validation equivalent locally; focused toolset/config tests passed 25/25 and the post-restart A2A card was healthy. The per-platform scoping in this PR is important so one platform's plugin names cannot mask another platform's typo.

PR #91213, opened 2026-08-21
feat(desktop): separate Bot Mode direct messages and groups

BkashJEE:
## Summary

Bot Mode currently presents direct-message and group-chat rows in one flat roster. This keeps the first-class group-row model introduced by #88738, while adding lightweight, independently collapsible **Direct Messages** and **Groups** sections so the two workflows remain easy to scan.

## What changed

- Adds separate Direct Messages and Groups roster headings without reverting #88738's flat, first-class group rows.
- Persists each section's disclosure state independently.
- Keeps Groups visible when Direct Messages are collapsed.
- Preserves attention visibility in collapsed section headers.
- Uses accessible disclosure controls and theme-native separation.
- Keeps search behavior useful by revealing matching content across both sections.

## Scope

This PR is intentionally limited to the Bot Mode roster source and focused tests. It excludes Selection Actions, mention-handle changes, Telegram routing, release artifacts, backups, and unrelated local work.

## Validation

- `node --check apps/desktop/src/plugins/hermes-bots/plugin.js`
- `node --test apps/desktop/src/plugins/hermes-bots/tests/roster-groups.test.mjs` — 11/11 passed
- `npm run check:test:plugins` — 352 passed, 0 failed
- `git diff --check`

`npm run check:lint` could not start in the clean worktree because `tsc` was unavailable; CI remains authoritative for the full lint/typecheck surface.

## Relationship to #88738

#88738 intentionally removed the old group section-header presentation and made each group a first-class row. This PR preserves that row model and ordering behavior. The headings here are presentation-level disclosure boundaries between direct-message rows and group rows, not a return to member sub-rows or the former group-detail section UI.

PR #97375, opened 2026-08-28
feat(desktop): persist collapsed Bot Mode roster sections

BkashJEE:
## Summary

Collapsing a gateway section or **Group chats** in the Bot Mode roster only holds for the current window. The set lives in pane state (`useState<Set<string>>`), so it is gone on the next restart — a user who keeps most gateways collapsed re-collapses them on every launch.

This moves the set to a `$collapsedRosterSections` atom in `bot-state.ts` — the leaf module Bot Mode's surfaces already share — persisted under `collapsed-roster-sections-v1`, following the `selected-roster-bot-v1` precedent directly above it.

Presentation state only: collapsing hides rows. It never detaches a gateway, opens or closes a chat, or touches bot metadata. Living in `bot-state` also means the set survives an unmount, not just a restart.

## What changed

- `bot-state.ts` — `$collapsedRosterSections`, `setRosterSectionCollapsed`, `hydrateCollapsedRosterSections`, `pruneCollapsedRosterSections`, and the storage key.
- `roster-pane.tsx` — reads the atom instead of local state; `toggleRosterSection` delegates to the setter; a prune effect keyed on the gateway id list.
- `plugin.tsx` — boot hydration alongside the existing selected-bot restore.

## Two details worth calling out

**Hydration races the user.** The boot read is async, so a click landing before storage answers would be silently undone when the older snapshot resolves. The reader captures a mutation generation up front and drops the payload if a toggle bumped it in the meantime. `selected-roster-bot-v1` needs no such guard (a single boot-time write), so copying that pattern as-is reintroduces the bug.

**Section ids embed connection ids.** They are `gateway:<connectionId>` and `group-chats`, so a deleted connection would otherwise leave its id in storage forever. The prune runs against the full gateway option list — never a search-filtered one — keeps the synthesized `legacy` and `all` bucket ids, and is a no-op on an empty list, which is what the roster looks like before sources have loaded.

The prune effect keys on the joined id list rather than the `gatewayOptions` array, which `rosterGatewayOptions` rebuilds on every render.

## Validation

- `vitest run src/plugins/hermes-bots/bot-state.roster-section-collapse.test.ts` — 10/10 passed (new file)
- `vitest run --project ui src/plugins/hermes-bots` — 58 files, 566 tests, 0 failed
- `npm run typecheck` — clean
- `npm run lint` — clean on all four touched files
- `npm run test:desktop:all` — packaged app + NSIS installer built clean

New tests cover the toggle/persist round trip, blank-id rejection, hydration de-duplication, non-array payloads, the click-during-hydration race, prune dropping dead gateways, prune keeping `legacy`/`all`, and prune refusing to write when there is nothing to drop or the option list is empty.

## Note on the rebase

This PR originally targeted the monolithic `plugin.js`. #96726 split that file into per-surface TypeScript modules while this was open, so the change has been reimplemented against the new layout rather than rebased — same behavior and the same two guards, now in `bot-state.ts` / `roster-pane.tsx` / `plugin.tsx` with a vitest suite in place of the old vm harness.

## Relationship to #91213

#91213 proposed Direct Messages / Groups sections with persisted disclosure state. Its sectioning has since been superseded — main groups bots per gateway and already renders Group chats as its own collapsible section. Persistence is the one part that did not land, so this PR carries just that across. #91213 is closed in favor of this.

### DIFF
--- a/apps/desktop/src/plugins/hermes-bots/bot-state.roster-section-collapse.test.ts
+++ b/apps/desktop/src/plugins/hermes-bots/bot-state.roster-section-collapse.test.ts
@@ -156,7 +156,9 @@ describe('pruning ids for gateways that are gone', () => {
 
     pruneCollapsedRosterSections([{ connectionId: 'alpha' }])
 
-    expect($collapsedRosterSections.get()).toEqual(['gateway:legacy', 'gateway:all'])
+    // Stored sorted, so the assertion reads in sorted order rather than the
+    // order the two were collapsed in.
+    expect($collapsedRosterSections.get()).toEqual(['gateway:all', 'gateway:legacy'])
   })
 
   it('is a no-op before sources load, so an empty list cannot wipe the set', async () => {
--- a/apps/desktop/src/plugins/hermes-bots/bot-state.roster-section-collapse.test.ts
+++ b/apps/desktop/src/plugins/hermes-bots/bot-state.roster-section-collapse.test.ts
@@ -1,0 +1,185 @@
+/**
+ * Collapsing a roster section is presentation state, but it is state the user
+ * SET — so it has to outlive the window. Two things make the persistence less
+ * trivial than the selected-bot key it sits beside:
+ *
+ * The boot read is async, so a click landing before storage answers would be
+ * silently undone when the older snapshot resolves. The reader carries the
+ * mutation generation it read at and drops a payload that lost the race.
+ *
+ * Section ids embed connection ids (`gateway:<connectionId>`), so a deleted
+ * connection would otherwise leave its id in storage forever. Pruning runs
+ * against the FULL gateway option list and must never fire on an empty one —
+ * that is what the roster looks like before sources have loaded, and treating
+ * it as "no gateways exist" would wipe every collapse on launch.
+ */
+
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const { writes } = vi.hoisted(() => ({ writes: [] as [string, unknown][] }))
+
+vi.mock('@hermes/plugin-sdk', async () => {
+  const { atom } = await import('nanostores')
+
+  return {
+    atom,
+    host: {
+      activeConnectionId: () => '',
+      state: {
+        connectionId: { get: () => 'source-a', listen: () => () => undefined },
+        profile: { get: () => 'default', listen: () => () => undefined }
+      }
+    },
+    queryClient: { invalidateQueries: vi.fn() },
+    useQuery: vi.fn(),
+    useValue: vi.fn()
+  }
+})
+
+vi.mock('./shared', () => ({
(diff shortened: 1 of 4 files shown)


Enough1122:

**Overall:** Persists collapsed Bot Mode roster sections across restarts with generation guard to avoid clobbering concurrent toggle.

**What it does**
- `apps/desktop/src/plugins/hermes-bots/plugin.js` adds `COLLAPSED_ROSTER_SECTIONS_KEY='collapsed-roster-sections-v1'` atom `$collapsedRosterSections`, `rosterSectionMutationGeneration` counter, `setRosterSectionCollapsed(id, collapsed)` (Set dedup, bumps generation, `pluginCtx.storage.set(key, list).catch(()=>undefined)`), `hydrateCollapsedRosterSections(value, expectedGeneration)` (if generation already bumped, user's click wins — does not overwrite newer state), and wiring into roster section render (collapsed ids are `gateway:<connectionId>` / `group-chats`). Presentation-only, never detaches gateway.

**Non-blocking notes**
- `.catch(()=>undefined)` swallows storage errors silently — consider debug log for persistence failures.
- Set→list conversion preserves insertion order but not sorted — stable enough; consider sorting for deterministic storage.

No blocking issues.

BkashJEE:
Thanks — both notes addressed or answered.

**Sorted storage:** taken, in e5ac8ebc8c. The id list is now sorted on write and on hydrate, so the same set of collapsed sections serializes identically regardless of click order, and a value written by an earlier build normalizes on first read.

**Silent `.catch(() => undefined)`:** leaving as-is deliberately. The neighbouring `selected-roster-bot-v1` writes in the same module swallow storage errors the same way, and both are presentation-only state — a failed write costs the user a collapsed section after restart, nothing more. Adding a log to this one write alone would make it the odd one out; if the swallowing is worth revisiting it should be for all the presentation-state writes together, not just this one.

One note for anyone reading the review above against the current diff: it ran before #96726 landed, so it describes the change as it was in `plugin.js`. That file has since been split into per-surface TypeScript modules and this PR was reimplemented against the new layout — the logic it reviewed now lives in `bot-state.ts` / `roster-pane.tsx` / `plugin.tsx`, with a vitest suite in place of the old vm harness. Both notes it raised still applied to the ported code.

teknium1:
Thanks for this. Roster organization landed on main via #101290 (user-made sections: create from the pane + menu or right-click → Move to section, drag-and-drop filing, rename/reorder/delete with Undo, nested inside each connection bucket, membership in the bot's ui_meta so it rides profile sync). We picked one mechanism for the four overlapping roster-organization PRs rather than layering them; if an ask of yours is not covered by sections (e.g. manual ordering within a section), please open a focused follow-up against current main. Closing in favor of the merged design.

PR #97376, opened 2026-08-28
fix(tools): relay the message_agent reply, not the turn's trailing narration

joshglenn:
## What does this PR do?

Fixes a Bot Mode agent-to-agent messaging bug: the reply relayed back to the
**sending** agent is not the text the receiving agent actually composed, but a
second, unrelated closing remark written afterward in the receiver's own
"talking to the user" voice.

Root cause is in `tools/bot_mode_dm.py::_run_delivery`. The completion
notification returned to the sender captures **whatever the target turn's final
stdout was** (`-Q` one-shot mode prints the turn's final assistant text). When
the receiving agent does the normal tool-calling thing — call `message_agent`
to reply, *then* add a short wrap-up after the tool result — that trailing
wrap-up (often addressed to the human, e.g. "…I'll relay it to you", "All good,
Josh") is what gets captured and relayed. The real, correctly-addressed reply —
the `message_agent` argument — is ded.

This approach reads the receiver's ack from its
per-profile `state.db` and relays *ack to the existing
stdout capture only when the receivno `message_agent`
call). The already-correct direct-ilain-text replies
are untouched.

## Related Issue

No separate issue filed — full reprate.db` evidence,
and root-cause analysis are includebody.

## Type of Change

- [x] 🐛 Bug fix (non-breaking chan

## Changes Made

- `tools/bot_mode_dm.py` — after th the target's own
  `message_agent` reply addressed b per-profile
  `state.db` and relay that insteaddout. Falls back to
  stdout on plain-text replies or aworse than before).
  Helpers: `_target_db_path`, `_senhigh_water`,
  `_peer_reply_after`.
- `tools/bot_mode_probe.py` — tightrotocol so a reply
  IS the `message_agent` call, withended after it in
  the same turn.

## How to Test

1. From a Bot Chat, `message_agent(age="ping 1")`.
2. Wait for the completion notifica
3. Query the target's Bot Chat sess
   ```sql
   SELECT id, role, content, tool_c
   WHERE session_id = '<target Bot  id DESC LIMIT 6;
4. Confirm the notification text noent of the
target's own message_agent tool calt text in
that turn.

Automated: 8 new cases in tests/toolay prefers the
message_agent payload over trailing narration; plain-text fallback; handle
normalization; since-id window). teest_bot_mode_probe.py (14) pass.

Checklist

Code

- [x] I've read the Contributing Gu
- [x] My commit messages follow Conls): …)
- [x] I searched for existing PRs tuplicate
- [x] My PR contains only changes r
- [x] I've run the affected suites:assed) + test_bot_mode_probe.py (14
passed)
- [x] I've added tests for my changools/test_bot_mode_dm.py)
- [x] I've tested on my platform: milicon)

Documentation & Housekeeping

- [x] I've updated relevant documentation (docstrings) — or N/A
- [x] I've updated cli-config.yaml. config keys — N/A (no config keys)
- [x] I've updated CONTRIBUTING.md/hitecture — N/A
- [x] I've considered cross-platforre sqlite3/pathlib; no
platform-specific primitives

### DIFF
--- a/tests/tools/test_bot_mode_dm.py
+++ b/tests/tools/test_bot_mode_dm.py
@@ -638,3 +638,116 @@ def test_dm_dir_rejects_precreated_symlink(tmp_path, monkeypatch):
 
     with pytest.raises(PermissionError, match="not a directory"):
         bot_mode_dm._dm_dir()
+
+
+# ── reply read-back: relay the message_agent payload, not trailing narration ──
+# Regression for the Bot-Mode "Track B" bug: _run_delivery captured the target
+# turn's final stdout (often a user-facing wrap-up) instead of the peer-addressed
+# message_agent reply the target actually composed. _peer_reply_after reads that
+# reply back from the target's own state.db so the right text is relayed.
+
+import sqlite3
+
+
+def _seed_target_db(path: Path, rows: list[dict]) -> None:
+    """Minimal `messages` table with the columns _peer_reply_after reads."""
+    conn = sqlite3.connect(str(path))
+    conn.execute(
+        "CREATE TABLE messages (id INTEGER PRIMARY KEY, session_id TEXT, "
+        "role TEXT, content TEXT, tool_calls TEXT, tool_name TEXT, timestamp REAL)"
+    )
+    for i, r in enumerate(rows, start=1):
+        conn.execute(
+            "INSERT INTO messages (id, session_id, role, content, tool_calls, "
+            "tool_name, timestamp) VALUES (?,?,?,?,?,?,?)",
+            (i, "sess-1", r["role"], r.get("content"), r.get("tool_calls"),
+             r.get("tool_name"), float(i)),
+        )
+    conn.commit()
+    conn.close()
+
+
+def _ma_call(target: str, message: str) -> str:
+    return json.dumps(
+        [{"function": {"name": "message_agent",
+                       "arguments": json.dumps({"target": target, "message": message})}}]
+    )
+
+
+def test_peer_reply_after_prefers_message_agent_over_trailing_narration(tmp_path):
+    db = tmp_path / "state.db"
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Fixes Bot-Mode "Track B" bug where DM reply relay captured trailing narration instead of `message_agent` payload.

**What it does**
- `tools/bot_mode_dm.py` adds `_peer_reply_after(session_id, since_row_id, db_path)` reading `state.db:messages` (tool_calls JSON) and preferring `message_agent` tool call's `message` argument over final assistant stdout. `_run_delivery` now calls it to relay peer's actual reply.
- `tests/tools/test_bot_mode_dm.py` seeds minimal `messages` table via `_seed_target_db`, helper `_ma_call(target, message)`, and tests `test_peer_reply_after_prefers_message_agent_over_trailing_narration` plus other relay cases.

**Non-blocking notes**
- SQLite read is read-only and filtered by `session_id` + `id > since` — correct scoping. Ensure `timestamp` float(i) in test matches production real timestamps (shape same).
- Wrap-up narration after `message_agent` tool still present in DB but ignored — intended.

teknium1:
Thanks @joshglenn for the careful trace. This is working as designed, though the design is easy to misread: message_agent is the one send path, so when the receiving bot replies via `message_agent` that reply already arrives at the sender as its own `Message from 🤖 <name> (@handle):` DM, and the completion notification carries the target turn's final text (e26d91dc11e). Reading the receiver's `message_agent` argument back out of its state.db and relaying it as the completion would deliver the same reply twice.
The stray narration you saw is the receiver ending its turn with prose after the tool call; a bare silence marker there is already blanked (`gateway/response_filters.py::is_intentional_silence_response`), so the lever is the protocol text, not a second relay path.
Closing as not-a-bug; happy to take a tight protocol-wording PR if the narration keeps showing up.

PR #97377, opened 2026-08-28
fix(cron): prevent false gateway-down warnings in gateway process

### LINKED ISSUE #97291: cronjob tool reports gateway_running:false from inside a live gateway (liveness probe false negative in-process)
## Summary

The `cronjob` model tool (and its `_gateway_liveness_notice` payload) reports `gateway_running: false` + "jobs will NOT fire until the gateway is started" **while the gateway is demonstrably running and firing jobs** — but only when the tool call executes inside the gateway process itself. The same probe chain returns True from any subprocess.

## Environment
- v0.20.5 (editable install, checkout SHA `58aebc72`), Linux, systemd unit `hermes-gateway.service`, default profile
- Gateway pid 2465630, up 40h+, `gateway_state.json` = `"gateway_state":"running"`, Telegram platform connected

## Reproduction
1. With the gateway running and jobs firing (executions.db shows fresh `claimed_at` rows every 15 min), issue a `cronjob list` from a chat session served by that gateway.
2. Result payload ends with:
   ```json
   "gateway_running": false,
   "warning": "The Hermes gateway is not running — these jobs are saved but will NOT fire until the gateway is started ..."
   ```
3. Run the identical chain from a subprocess on the same box (same venv python, same env, same cwd):
   ```python
   from tools.cronjob_tools import _gateway_liveness_notice   # -> {'gateway_running': True}
   from hermes_cli.cron import _builtin_gateway_liveness      # -> True
   from hermes_cli.gateway import find_gateway_pids           # -> [2465630]
   from gateway.status import get_running_pid                 # -> 2465630
   ```
   All True. Repeated across sessions/hours; the in-gateway result is False every time, so it is deterministic, not a race window.

## Impact
- The warning is factually wrong (jobs fire — the ticker is alive) and trains users to ignore a load-bearing warning whose purpose is catching the "jobs silently not firing" failure mode (#51038).
- Agents acting on the notice may try to (re)start the gateway or tell users monitoring is down.

## Suspects (not fully isolated)
The chain is `_gateway_liveness_notice` → `_builtin_gateway_liveness` → provider==builtin → `find_gateway_pids()` → three paths: `gateway.status.get_running_pid()` (PID-file + runtime-lock check), `_get_service_pids` (systemd), `_scan_gateway_pids` (proc scan). Inside the gateway process all three appear to come up empty. Candidates: the runtime-lock self-check in `is_gateway_runtime_lock_active` (`gateway/status.py` ~L1003: a probe running in the lock-owning process takes the `_gateway_lock_handle` fast path or succeeds at acquiring its own lock via fcntl semantics), or PID-record/self-PID filtering in `_record_matches_live_gateway_pid`/leaf-PID drop logic dropping the gateway's own PID.

## Suggested direction
The probe asking "is a gateway alive?" must never return a false negative **when called from inside a live gateway**: if `os.getpid()` matches the running-gateway PID record (or the process holds the runtime lock handle / `gateway_state.json` says running with our own start_time), liveness is proven outright — short-circuit to True before the scan paths. Cheap, self-contained, and kills the whole class.

Happy to test a patch — we run a git checkout and can verify in-place.

trevorswanson:
## Summary

The cron liveness warning could report `gateway_running: false` from a model-tool call executing inside a live gateway process. This patch makes the shared resolver recognize the gateway process itself as authoritative for the active Hermes profile, while retaining the tri-state `None` behavior for probe errors.

## Changes

- Use `resolve_gateway_liveness()` from the builtin cron adapter instead of the weaker `find_gateway_pids()` heuristic.
- Return unknown rather than a confident stopped result when the resolver reports a probe error.
- Short-circuit the canonical resolver to the current PID when its command line is `gateway run` and the requested profile is the active Hermes home.
- Update cron regression mocks and add an in-process/profile-scoping regression test.

## Validation

- `pytest -q tests/gateway/test_status.py tests/cron/test_87033_cronjob_gateway_liveness.py`
- Result: **77 passed, 2 skipped**
- `python3 -m compileall` and `git diff --check` pass.

Fixes #97291

### DIFF
--- a/gateway/status.py
+++ b/gateway/status.py
@@ -1349,6 +1349,16 @@ def resolve_gateway_liveness(
     monkeypatch seam in the test-suite keeps working; production callers
     leave them ``None`` and get this module's implementations.
     """
+    # A model tool can execute in the gateway's own process. In that context
+    # process-table scans and the gateway's self-held runtime lock can both
+    # produce a false negative even though the caller is the gateway runtime.
+    # Treat the process identity as authoritative, but only for the active
+    # Hermes home so a multiplexed gateway cannot answer for another profile.
+    active_home = get_hermes_home().resolve()
+    requested_home = profile_dir.resolve() if profile_dir is not None else active_home
+    if requested_home == active_home and looks_like_gateway_command_line(" ".join(sys.argv)):
+        return GatewayLiveness(running=True, pid=os.getpid(), source="self")
+
     _pid_probe = pid_probe or (
         get_running_pid_cached if use_cache else get_running_pid
     )
--- a/hermes_cli/cron.py
+++ b/hermes_cli/cron.py
@@ -78,9 +78,16 @@ def _builtin_gateway_liveness() -> Optional[bool]:
     try:
         if _active_cron_provider_name() != "builtin":
             return True  # external provider fires jobs without the gateway
-        from hermes_cli.gateway import find_gateway_pids
+        from gateway.status import resolve_gateway_liveness
+        from hermes_constants import get_hermes_home
 
-        return bool(find_gateway_pids())
+        result = resolve_gateway_liveness(profile_dir=get_hermes_home())
+        # ``False`` plus a probe error means "unknown", not "stopped". The
+        # cron warning is user-facing, so it must never turn an unreadable
+        # status source into a confident outage claim.
+        if result.probe_error and not result.running:
+            return None
+        return result.running
     except Exception:
         return None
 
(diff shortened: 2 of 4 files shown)


trevorswanson:
Closing as superseded by upstream main. Commit b71c3cc13c (Aug 27) already adds the runtime-lock-first in-process fast path for builtin cron liveness, and 1e5fb70fb5 extends it to cron status. This PR is redundant rather than an additional fix.

PR #97380, opened 2026-08-28
fix(updater): stop Windows hand-off updates leaving Desktop stale

### LINKED ISSUE #97343: [Bug]: Windows shim hand-off skips desktop rebuild after hermes update
### Bug Description

On Windows (git install), `hermes update` run from a PowerShell-launched `hermes.exe` takes the Windows shim hand-off path: the updater spawns a venv-Python child (`HERMES_UPDATE_REEXEC=1`) to finish the dependency sync, because hermes.exe cannot replace itself while running.

That child takes the "Already up to date" / dependency-repair completion path inside `_cmd_update_impl` (hermes_cli/update_cmd.py), which prints `✓ Dependencies repaired!` → `✓ Update complete!` and returns — **without ever calling `_rebuild_desktop_after_update()`**.

The main commits-pulled path calls `_rebuild_desktop_after_update()` (update_cmd.py:2129 and :8454), which runs the content-hash-stamp check (`_desktop_build_needed`, hermes_cli/main.py:6842) and rebuilds via `hermes desktop --build-only` when desktop source changed. The hand-off/repair completion path omits this step entirely (the same omission exists in `_repair_node_deps_on_current_checkout`, update_cmd.py:3707).

Net effect: every Windows user who runs `hermes update` from a console (the normal case) gets a **silently stale packaged desktop app** whenever the pulled commits touch `apps/desktop/`. The update output shows success, with no `→ Checking if desktop app needs rebuilding...` line and no warning that the desktop binary is now older than the source.

This is the same aftermath class as #70337/#44225 (desktop app wiped/stale after update), but it occurs on the *normal* update path even when the ZIP fallback is never taken and `apps/desktop/release/` is preserved.

Related but distinct: #86443 covers the rebuild running and failing (clean-before-deps destroys the app, failure suppressed); #41178 (closed via #42100) covered file-lock failures during rebuild. This issue is the third failure mode: on the HERMES_UPDATE_REEXEC hand-off/repair path the rebuild is never invoked at all, so neither of those fixes applies — the stamp check and hermes desktop --build-only are simply never reached.

### Steps to Reproduce

1. Windows, git install. Packaged desktop app present (apps/desktop/release/win-unpacked/Hermes.exe + current desktop-build-stamp.json).
2. From PowerShell (not inside the desktop app), run `hermes update` when upstream has commits touching apps/desktop/.
3. Updater pulls, hits the Windows shim self-lock, prints "Windows: hermes.exe cannot replace itself while it runs; finishing the dependency install under the venv Python", spawns the hand-off child, and that child completes the update.
4. Full output ends with `✓ Dependencies repaired!` → `✓ Update complete!` — no `→ Checking if desktop app needs rebuilding...` line anywhere.
5. Hermes.exe mtime and desktop-build-stamp.json still predate the pulled commits; `git log --oneline <old>..HEAD -- apps/desktop` shows desktop source changed.

Verified 2026-08-28: update to 6dcebea7fc5 pulled 210 commits (82 tracked files changed under apps/desktop/); update completed with no rebuild attempted; binary stayed at the pre-update build. Adding the missing `_rebuild_desktop_after_update()` call to the hand-off completion path (local patch) makes the same flow run the stamp-gated rebuild correctly.

### Expected Behavior

The hand-off/repair completion path should run the same desktop-rebuild tail as the main commits-pulled path: check the content-hash stamp, and if desktop source changed, rebuild the packaged app (`hermes desktop --build-only`) before printing "✓ Update complete!" — or at minimum print a hard warning that the desktop binary is stale and how to rebuild it.

### Actual Behavior

The hand-off child prints `✓ Dependencies repaired!` → `✓ Update complete!` and returns. No desktop check, no rebuild, no warning. Hermes.exe remains on the pre-update build while the source tree (including apps/desktop/) is now newer — silent divergence between packaged app and code.

### Affected Component

CLI (interactive chat)

N/A (CLI only)

### Debug Report

```shell
Report:      https://paste.rs/0xQFj
agent.log:   https://paste.rs/EUUxD
gateway.log:   https://paste.rs/ZvqXJ
gui.log:     https://paste.rs/wt21L
desktop.log: https://dpaste.com/F9MWGF3PJ
```

### Operating System

Windows 10 (10.0.19045) AMD64

### Python Version

3.11.15

### Hermes Version

v0.20.6 (2026.8.27) @ eff97a8a, git install

```shell
update.log (2026-08-28 run, hand-off path — note the ABSENCE of any desktop-rebuild line):

=== hermes update started 2026-08-27T20:35:06 ===
⚕ Updating Hermes Agent...
→ Stopping Windows gateway process(es) before updating Hermes...
  → Force-stopped 1 gateway process(es)
  → Stopped 1 gateway process(es) without profile mapping
→ Fetching updates...
→ Finishing the dependency install handed off by hermes.exe...
→ Refreshing 20 active lazy backend(s)...
  ✓ 20 already current
✓ Dependencies repaired!
→ Checking configuration for new options...
  ✓ Configuration is up to date
✓ Update complete! [main @ 6dcebea7fc5]

For contrast, a non-hand-off run in the same log DOES include:
  → Checking if desktop app needs rebuilding...
  ✓ Desktop packaged app ready: ...\release\win-unpacked\Hermes.exe
  ✓ Desktop app up to date
```

`_rebuild_desktop_after_update()` (update_cmd.py:7129) — the only caller of the stamp check `_desktop_build_needed()` (main.py:6842) in the update flow — is invoked only from the main commits-pulled path (update_cmd.py:2129 and :8454). The Windows shim hand-off child (spawned by `_reexec_dependency_sync_off_windows_shim`, main.py:9288, via HERMES_UPDATE_REEXEC=1) re-enters `_cmd_update_impl` with the checkout already current, takes the `commit_count == 0` repair branch (~line 7915), finishes the dependency sync, prints "✓ Update complete!" and returns — never reaching the rebuild call. `_repair_node_deps_on_current_checkout` (update_cmd.py:3707), the non-hand-off variant of the same branch, has the identical omission.

The hand-off fires on essentially every console-launched `hermes update` on Windows (hermes.exe holds itself open), so the main path's rebuild step is effectively dead for CLI users on Windows.

`had_desktop_app_before_update` is captured pre-branch (update_cmd.py:7522) and is in scope at the insertion point (AST-verified on the live checkout).

### Proposed Fix (optional)

In update_cmd.py, hand-off/repair completion branch (inside `_cmd_update_impl`, "Already up to date" path), insert before `_print_update_completion(...)`:

    _rebuild_desktop_after_update(
        _m().PROJECT_ROOT / "apps" / "desktop",
        had_desktop_app_before_update=had_desktop_app_before_update,
    )

Verified locally: with this 9-line patch, the identical update flow ran the stamp-gated rebuild ("→ Building desktop packaged app..." → "✓ Desktop packaged app ready" → stamp rewritten) instead of silently completing with a stale binary. Same call should be added to `_repair_node_deps_on_current_checkout` for the non-hand-off repair path.

fangliquanflq:
## What does this PR do?

Windows shim hand-off updates could finish successfully while leaving the packaged Desktop app older than the updated source. This change gives both current-checkout repair paths the same stamp-gated Desktop rebuild tail as a normal pulled update and withholds a success completion when that rebuild fails.

### Symptom

After `hermes update` hands dependency repair from `hermes.exe` to venv Python, the child can print `Dependencies repaired!` and `Update complete!` without printing the Desktop rebuild check. The existing packaged app and build stamp remain stale after Desktop source changes.

### Impact

Console-launched Windows updates can report success while users continue running an older packaged Desktop build.

### Bug Cause

**Trigger:** `hermes_cli/update_cmd.py` current-checkout branch in `_cmd_update_impl` when `handed_off_sync` is true or `commit_count == 0`.

**Causal chain:**
1. The first updater process pulls source changes, then hands dependency installation to a venv-Python child because the running Windows shim cannot replace itself.
2. The child sees the checkout already at the fetched tip and enters the current-checkout repair branch.
3. That branch repaired dependencies and printed completion without calling `_rebuild_desktop_after_update`, so Desktop stamp validation never ran.

**Why it is wrong:** A current git tip does not mean ignored packaged Desktop artifacts match that tip. Completion must account for the Desktop rebuild result.

**Working sibling / contrast:** The normal git-pull and ZIP update paths already call `_rebuild_desktop_after_update` before their final update summary.

**Ruled out:** This is not a Desktop stamp calculation failure. The affected branch never invoked the existing stamp check at all.

### Fix

Add one shared current-checkout completion helper that runs the Desktop rebuild check, applies config migration, reports partial completion on rebuild failure, and writes the gateway exit marker when applicable. Route both successful Python hand-off repair and ordinary Node repair through that helper.

## Related Issue

Fixes #97343

## Type of Change

## Changes Made

- `hermes_cli/update_cmd.py` - run the Desktop rebuild tail before successful current-checkout completion.
- `tests/hermes_cli/test_cmd_update.py` - cover successful rebuild ordering and failure reporting.

## How to Test

1. Run the focused current-checkout completion regressions.
2. Run the updater branch-fallback and Desktop rebuild test classes.

```bash
scripts/run_tests.sh tests/hermes_cli/test_cmd_update.py -k 'current_checkout_completion' -q
scripts/run_tests.sh tests/hermes_cli/test_cmd_update.py -k 'TestNodeRuntimeNpmResolution' -q
scripts/run_tests.sh tests/hermes_cli/test_cmd_update.py -k 'TestCmdUpdateBranchFallback' -q
```

### Code

- [x] I've run the relevant updater tests and all selected tests pass
- [x] I've tested the unit-test path on Windows 11

- [x] Documentation updates are N/A
- [x] `cli-config.yaml.example` updates are N/A
- [x] `CONTRIBUTING.md` and `AGENTS.md` updates are N/A
- [x] I've considered cross-platform impact; the shared helper preserves existing platform-gated Desktop behavior
- [x] Tool description and schema updates are N/A


liuhao1024:
Independent review — I had started on this from the issue and am standing down in favor of this PR; the coverage here is complete and the shape is right. Two notes from cross-checking against `upstream/main`:

**1. The new required kwarg breaks 4 existing tests.** `_repair_node_deps_on_current_checkout` now takes `had_desktop_app_before_update: bool` as a *required* keyword-only argument, but these existing callers in the test suite don't pass it and will raise `TypeError` on CI:

- `tests/hermes_cli/test_update_current_node_repair.py:24` and `:40` (`test_current_checkout_repairs_failed_node_deps`, `test_current_checkout_healthy_node_deps_reports_up_to_date`)
- `tests/hermes_cli/test_update_config_migration_on_current_checkout.py:33` and `:56` (`test_repair_node_deps_runs_config_migration_on_version_bump`, `test_repair_node_deps_up_to_date_config`)

Cheapest fix: default it to `False` and skip the rebuild inside `_finish_current_checkout_update` when `had_desktop_app_before_update=False` and no desktop app is present — `_rebuild_desktop_after_update` already self-gates on `_desktop_app_present`, so a `False` default costs nothing on desktop-less installs. (Alternatively update the 4 call sites.)

**2. Wiring-level test blind spot (non-blocking).** Both new tests call `_finish_current_checkout_update` directly, so they stay green even if the `_cmd_update_impl` call site were accidentally removed — the original bug was exactly a missing call, so a test that drives the real function would pin the wiring. If you want it: drive `_cmd_update_impl` with `HERMES_UPDATE_REEXEC=1` set, git plumbing stubbed to report `commit_count == 0` on `main` (same approach `test_update_venv_health.py` uses to reach this branch's earlier phases), and assert `update_cmd._rebuild_desktop_after_update` gets called with `PROJECT_ROOT / "apps" / "desktop"`. Happy to paste a working version if useful.

fangliquanflq:
Thanks for the careful cross-check. I addressed both points:

1. **Required keyword argument — valid, already addressed.** The four existing `_repair_node_deps_on_current_checkout` test callers now pass `had_desktop_app_before_update=False` explicitly, and the healthy paths mock the Desktop rebuild result where needed (commit `3e210005b7`).
2. **Wiring-level coverage — valid and now added.** `test_current_checkout_handoff_wires_desktop_rebuild` executes the real `_cmd_update_impl` no-commit branch with `HERMES_UPDATE_REEXEC=1`, then verifies that the Desktop rebuild receives `PROJECT_ROOT / "apps" / "desktop"`, preserves the pre-update Desktop state, and runs before successful completion (commit `97b3080884`).

I also merged current upstream `main`; the only conflict was resolved by preserving its customizable current-checkout completion message and forwarding that value through the shared rebuild tail.

Validation:
- `scripts/run_tests.sh tests/hermes_cli/test_cmd_update.py -k "current_checkout_completion or TestNodeRuntimeNpmResolution or TestCmdUpdateBranchFallback" -q` — 10 passed
- `scripts/run_tests.sh tests/hermes_cli/test_update_config_migration_on_current_checkout.py tests/hermes_cli/test_update_current_node_repair.py tests/hermes_cli/test_update_venv_health.py -q` — 13 passed
- Ruff on all changed files, `git diff --check`, and the Windows footgun scan passed during independent review.

egilewski:
looks mergeable

The PR routes current-checkout repair and Windows hand-off completion through the existing Desktop rebuild gate, preserves failure reporting, and introduces no new untrusted input or privilege boundary. Offline cross-revision checks passed; no source-backed security finding.

Security evidence:
- trust boundary: An operator or authorized gateway invokes the updater. Remote update contents and local checkout/dependency/Desktop metadata are the potential untrusted sources inside the existing update trust boundary; the changed tail builds the local Desktop artifact and writes the gateway completion marker through existing mechanisms.
- source/sink/invariant: The changed callers reach one shared tail only after dependency health succeeds. It passes a fixed project-relative Desktop path and a list-form Python command to the existing rebuild gate; a failed build withholds success and records failure in gateway mode, while a successful build permits the supplied completion message.
- current-main reproduction: The current-main tree was extracted into scratch and its updater regression selectors ran with HERSEC_REVIEW_PYTHON, -B, scratch-only HOME/TMPDIR, and pytest cache disabled: 12 tests passed. The baseline repair path completed configuration migration and its existing completion behavior without the new Desktop-tail call.
- PR-head or patch-replay validation: The reviewed head was patch-replayed onto current main in the bound checkout. The four updater selector files, including the new success/failure and hand-off wiring controls, passed: 54 tests. The staged diff check was clean.
- positive/negative cases: Positive control: the shared tail calls the completion callback when the rebuild gate returns true. Negative control: a false rebuild result suppresses completion and emits the partial-update summary. Existing node-failure early return remains separate and does not claim a successful repair.
- residual bypass search: All implementation and test call sites for the current-checkout repair helper were enumerated. The normal pull and ZIP paths already used the same Desktop rebuild gate; the PR adds only the two current-checkout repair callers. The local updater's pre-existing execution of trusted checkout package scripts is unchanged in kind.
- reviewer validation: CodeRabbit completed with no findings in the clean-pass flow.

Not checked:
- native_windows_execution

liyangbing:
<p>This fixes a subtle but user-visible updater contract: a Windows shim hand-off can repair the venv and still leave the packaged Desktop artifact stale. Reusing the same completion tail for current-checkout repair, gating the success message on the Desktop rebuild result, and writing the gateway exit marker from that result are the right boundaries.</p>

<h3>Suggested contract</h3>

<ul>
<li>Make every current-checkout completion entry point go through the shared tail exactly once, including Python hand-off repair, ordinary Node repair, gateway mode, and the no-commit path; add an integration test at <code>_cmd_update_impl</code> rather than only testing the helper in isolation.</li>
<li>Keep completion truthful across the process boundary. The hand-off child, parent shim, gateway supervisor, and CLI output must all agree that a failed Desktop rebuild is partial/failed, never a successful update.</li>
<li>Preserve resumability after a Desktop build failure: retain the fetched source and dependency state, report the stamp/build failure reason and affected path, and make a rerun retry the build without incorrectly repeating or skipping migrations.</li>
<li>Define ordering and failure semantics for config migration versus Desktop rebuild. If migration succeeds but the build fails, the persisted state and next-run behavior must remain safe and the summary must identify both outcomes separately.</li>
<li>Keep the pre-update Desktop identity and stamp evidence tied to the same checkout/profile. A missing app before update, a source-only update, and an already-stale packaged app must not be conflated.</li>
<li>Ensure the gateway exit marker is written for every relevant failure, is atomic/complete, and cannot be mistaken for a stale marker from a previous update or a child that exited before writing it.</li>
</ul>

<p>The added tests cover helper ordering and success suppression. The highest-value missing witness is a Windows-shaped hand-off test that runs the child/current-checkout branch, injects a successful dependency repair plus a failed Desktop rebuild, and asserts the parent/gateway receives the failure while no <code>Update complete!</code> text is emitted. Also assert the successful branch invokes the rebuild before migration/completion in the same way as a pulled or ZIP update.</p>

<h3>Regression matrix</h3>

<ul>
<li>normal git-pull, ZIP fallback, current checkout with zero commits, and Python hand-off repair</li>
<li>Desktop present before update, absent before update, stale stamp, unchanged source, and changed Desktop source</li>
<li>successful rebuild, build failure, missing toolchain, permission/ACL denial, cancellation, and interrupted child</li>
<li>config migration success/failure before and after Desktop failure, with resumable rerun</li>
<li>gateway mode, CLI mode, parent shim, child Python process, stale/missing/partial exit marker, and nonzero exit propagation</li>
<li>success text, partial summary, stamp identity, build logs, fetched revision, and rollback/retry evidence</li>
</ul>

<p>Evidence reviewed: <code>hermes_cli/update_cmd.py</code>, <code>tests/hermes_cli/test_cmd_update.py</code>, and the related current-checkout/health tests in PR head <code>97b30808844ae8a1d16efd53f0693f9fbad5a5aa</code>.</p>

fangliquanflq:
Thanks for the detailed contract review. I treated the missing end-to-end failure witness as valid actionable feedback and strengthened the real wiring test in `44359e9669e`.

- `test_current_checkout_handoff_reports_desktop_rebuild_result` now drives `_cmd_update_impl` with `HERMES_UPDATE_REEXEC=1` in gateway mode for both Desktop rebuild outcomes.
- The success case proves the shared tail runs exactly once in the order `rebuild -> migration -> completion -> gateway success` and preserves the captured pre-update Desktop state/path.
- The failure case proves the same hand-off branch emits the partial-update summary, never emits `Update complete!`, and propagates `False` to the gateway exit marker.

The remaining contract points do not require additional production changes in this PR:

- Both successful current-checkout entry points already converge on `_finish_current_checkout_update`; normal pull and ZIP paths already use `_rebuild_desktop_after_update`.
- Rebuild failure leaves the fetched checkout/dependency state intact, does not record a successful build, prints the retry command and update-log path, and config migration remains an independently reported step after the rebuild attempt.
- `had_desktop_app_before_update` is captured once from the same `PROJECT_ROOT` before update-path mutation and is forwarded unchanged to the stamp-gated rebuild helper, preserving the absent/present distinction.
- Gateway launch removes the previous `.update_exit_code` before spawning (`gateway/slash_commands.py`), while the watcher treats nonzero/timeout as failure. Broader atomic-marker and interrupted-supervisor redesign is pre-existing gateway infrastructure rather than behavior introduced by this PR.

Validation:
- `scripts/run_tests.sh tests/hermes_cli/test_update_venv_health.py tests/hermes_cli/test_update_config_migration_on_current_checkout.py tests/hermes_cli/test_update_current_node_repair.py -q` — 14 passed
- `scripts/run_tests.sh tests/hermes_cli/test_cmd_update.py -k "current_checkout_completion or TestNodeRuntimeNpmResolution or TestCmdUpdateBranchFallback" -q` — 10 passed
- Ruff and `git diff --check` passed
- Independent scan-only review verdict: perfect

teknium1:
Thanks @fangliquanflq — closing this one per maintainer decision. The Windows update/hand-off path has moved substantially this week (fail-closed taskkill identity guard #99558, self-lock deferral resume #99525, stash-restore validation #99555), and we don't want another layer on the hand-off sequence right now. If Desktop-stale-after-update reports persist on current main, we'll revisit with fresh evidence.

PR #54978, opened 2026-06-29
fix(steer): only inject /steer into a running turn; confirm busy-input steers

counterposition:
## TL;DR

Makes TUI/Desktop steering truthful and lossless across all session states:

1. An idle session can retain its agent object. `/steer` previously treated that stale agent as a live turn, queued text for an unrelated future turn, and claimed it would arrive after the next tool call. Live injection now requires `session["running"]`.
2. A newly submitted turn has the inverse startup window: `running` is already true while deferred agent construction is still in flight. Desktop rejects `send` directives while busy, so returning one there drops the text. On current `main`, this PR now uses the gateway's existing `queued_prompt` path for any running turn that cannot accept the steer, and returns an `exec` acknowledgement.
3. Ink busy-input steering now prints a success acknowledgement instead of making an accepted steer look like swallowed input.

## Behavior

- **Running + steer accepted:** inject after the next tool call and return an `exec` confirmation.
- **Running + agent still starting / steer rejected / steer failed:** retain the text server-side as the next turn and return an `exec` explanation. The queue operation is atomic with turn teardown, so the message cannot miss the drain edge.
- **Idle:** return `send` so the client submits it as a normal next turn. A retained idle agent is never steered.
- **Blank input:** return the `/steer <prompt>` usage error.

The server-side `command.dispatch` path is used by clients such as Electron Desktop. Ink's slash command is client-side, while busy-input steer uses `session.steer` directly.

## Why use `queued_prompt`

Recent `main` added a server-owned next-turn queue for mid-turn submissions, including client-transport pinning and post-turn draining. Reusing it avoids a second pending-message mechanism and fixes the startup race without relying on a busy client to resubmit.

## Tests

`tests/tui_gateway/test_steer_command.py` covers:

- accepted live steer and confirmation;
- trimming;
- idle retained-agent fallback without calling `steer()`;
- deferred-build startup race (running with no agent);
- rejected and raised steers queued server-side;
- idle no-agent fallback;
- empty and whitespace-only usage errors.

Validation:

```
python -m pytest tests/tui_gateway/test_steer_command.py -q
# 9 passed

python -m pytest \
  tests/test_tui_gateway_server.py::test_session_steer_calls_agent_steer_when_agent_supports_it \
  tests/test_tui_gateway_server.py::test_interrupt_drops_queued_prompt_for_session -q
# 2 passed

cd ui-tui && npm run typecheck
```

## Scope / safety

- One server handler, one Ink acknowledgement line, and focused regression tests.
[...]

PR #97381, opened 2026-08-28
fix(gateway): gate /steer on a live turn in command.dispatch

### LINKED ISSUE #97371: [Bug]: /steer - silently discards messages when no turn is running
### Bug Description

When using `/steer` in Hermes Desktop after the current agent turn has already finished, the message can be silently discarded.

The message is not added to the chat or session history, is not available through the Up Arrow composer history, and does not appear to be preserved in the prompt context or prompt cache.

The user receives no clear indication that the message was not processed. As a result, important user input can be lost without any warning.

### Steps to Reproduce

1. Open Hermes Desktop.
2. Start a task that takes long enough to allow mid-run interaction.
3. Wait until the current agent turn has finished.
4. Send a message using `/steer`, for example:
   `/steer Please do not change the configuration yet`
5. Observe that the `/steer` message does not appear in the chat history.
6. Press the Up Arrow key in the composer and observe that the message is not available there.
7. Check the session history and observe that the message has not been persisted.

The issue appears to be timing-sensitive: `/steer` works as a mid-run command while a turn is active, but may be silently discarded when the turn has just completed.

### Expected Behavior

`/steer` should never silently lose user input.

When a turn is still running, the current mid-run `/steer` behavior should be preserved.

When no turn is running, Hermes should automatically queue `/steer <message>` as the next prompt, similar to `/queue <message>`. The message should be persisted in the session history and remain available through the composer history.

Hermes should also show a clear confirmation, for example:

`No turn is currently running. The /steer message was queued as the next prompt.`

If automatic queuing is not possible, Hermes should explicitly reject the command and clearly explain that no active turn exists and that the message was not processed. Under no circumstances should the message be silently discarded.

### Actual Behavior

When `/steer` is sent after the current agent turn has finished, the message may be silently discarded.

The message does not appear in the chat or session history and cannot be recovered with the Up Arrow key in the composer. It is also not available through the persisted session history or prompt context.

The user receives no clear indication that no active turn existed, that `/steer` was not applied, or that the message was not queued for later processing.

The behavior is timing-sensitive: `/steer` works as a mid-run command while a turn is active, but may disappear when sent immediately after the turn has completed.

### Affected Component

CLI (interactive chat)

_No response_

### Debug Report

```shell
The issue is cross-platform and affects both Hermes Desktop and TUI.

It is reproducible as a timing-related input-loss problem with `/steer`: when the current agent turn has already finished, the `/steer` message may disappear instead of being persisted or queued.

The issue has been observed on macOS Tahoe 26.6.2 (25G83) and is not believed to be specific to macOS. It may also occur on other supported operating systems.

No full debug logs are attached because they are not necessary to describe this issue and may contain unrelated system information.
```

### Operating System

macOS Tahoe 26.6.2 (25G83)

### Python Version

_No response_

### Hermes Version

_No response_

```shell

```

_No response_

### Proposed Fix (optional)

_No response_

liuhao1024:
## What does this PR do?

`/steer` sent after a turn had already finished was silently discarded on the Desktop path. Root cause: Desktop routes `/steer` through `slash.exec` → `command.dispatch`, and the dispatch branch called `agent.steer(arg)` whenever an agent object existed — without checking `session["running"]`. `AIAgent.steer()` has no notion of a live turn: it stashes the text in `_pending_steer` and returns `True` unconditionally. On an idle session no drain hook ever fires (there is no next tool call), so the message sat unread while the UI had just reported "⏩ Steer queued". The turn-end leftover requeue (`result["pending_steer"]` in `_run_prompt_submit`'s tail) only captures steers that arrived *while* the turn was live, so it cannot cover this window.

Every other entry point already gates on liveness — the classic CLI checks `self._agent_running` (cli.py), the Ink TUI checks `ctx.ui.busy` before calling `session.steer`, and `_busy_submit` reads `session["running"]` under `history_lock` before steering. This PR adds the same gate to `command.dispatch`: read `session["running"]` under `history_lock` (mirroring the `/retry` branch in the same handler) and only steer a live session. An idle session falls through to the existing send fallback — `{"type": "send", "message": arg}` — which is exactly what `/queue` returns from the same handler, so the message is submitted as the next prompt instead of being lost.

The residual race (session observed running, turn settles milliseconds later) is the same window the other entry points accept, and it is covered by the turn-end leftover requeue.

## Related Issue

Fixes #97371

## Type of Change

## Changes Made

- `tui_gateway/methods_tools.py` — `command.dispatch` `/steer` branch: read `session["running"]` under `history_lock` before calling `agent.steer()`; an idle session (or a steer-less agent, as before) falls through to the `send` fallback.
- `tests/test_tui_gateway_server.py` — two regression tests: an idle session with a steer-capable agent must return `{"type": "send"}` and must NOT stash a steer; a running session still steers and reports "Steer queued".

## How to Test

1. `uv run --extra dev python -m pytest tests/test_tui_gateway_server.py -k command_dispatch_steer -q`
   - Observed result: 2 passed — idle session returns `type=send` with the payload and `agent.steer()` is never called; running session steers.
2. `uv run --extra dev python -m pytest tests/test_tui_gateway_server.py tests/test_tui_gateway_queue_on_busy.py -q`
   - Observed result: 649 passed (full tui_gateway suites, no regressions).
3. Manual (Desktop, pre-fix behavior from the issue): let a turn finish, then send `/steer <text>` — the message vanished. With the fix, the same command submits the text as the next prompt, identical to `/queue <text>`.

### Code

### DIFF
--- a/tests/test_tui_gateway_server.py
+++ b/tests/test_tui_gateway_server.py
@@ -11139,6 +11139,76 @@ def test_snapshot_restore_is_blocked_from_tui_worker():
     )
 
 
+def test_command_dispatch_steer_falls_back_to_send_when_session_idle(monkeypatch):
+    """An idle session must not accept a /steer (#97371).
+
+    AIAgent.steer() parks the text in _pending_steer regardless of whether a
+    turn is live; with no running turn no drain hook ever fires, so the
+    message was silently discarded right after the UI reported "Steer
+    queued". The dispatcher must gate on session["running"] and return the
+    send fallback instead — the same next-turn semantics as /queue.
+    """
+    monkeypatch.setattr(server, "_load_cfg", lambda: {})
+    calls = {}
+
+    class _Agent:
+        def steer(self, text):
+            calls["steer_text"] = text
+            return True
+
+    server._sessions["sid"] = _session(agent=_Agent())  # running=False
+    try:
+        resp = server.handle_request(
+            {
+                "id": "1",
+                "method": "command.dispatch",
+                "params": {
+                    "name": "steer",
+                    "arg": "do not change the config yet",
+                    "session_id": "sid",
+                },
+            }
+        )
+    finally:
+        server._sessions.pop("sid", None)
+
+    assert "result" in resp, resp
+    assert resp["result"]["type"] == "send"
+    assert resp["result"]["message"] == "do not change the config yet"
+    assert "steer_text" not in calls  # idle session must not stash a steer
+
+
+def test_command_dispatch_steer_injects_when_turn_running(monkeypatch):
+    monkeypatch.setattr(server, "_load_cfg", lambda: {})
+    calls = {}
+
+    class _Agent:
+        def steer(self, text):
+            calls["steer_text"] = text
+            return True
+
+    server._sessions["sid"] = _session(agent=_Agent(), running=True)
+    try:
+        resp = server.handle_request(
+            {
+                "id": "1",
+                "method": "command.dispatch",
+                "params": {
+                    "name": "steer",
+                    "arg": "also check auth.log",
+                    "session_id": "sid",
+                },
+            }
+        )
+    finally:
+        server._sessions.pop("sid", None)
+
+    assert "result" in resp, resp
+    assert resp["result"]["type"] == "exec"
+    assert "Steer queued" in resp["result"]["output"]
+    assert calls["steer_text"] == "also check auth.log"
+
+
 def test_command_dispatch_exec_nonzero_surfaces_error(monkeypatch):
     monkeypatch.setattr(
         server,
--- a/tui_gateway/methods_tools.py
+++ b/tui_gateway/methods_tools.py
@@ -798,7 +798,20 @@ def _(rid, params: dict) -> dict:
         if not arg:
             return _err(rid, 4004, "usage: /steer <prompt>")
         agent = session.get("agent") if session else None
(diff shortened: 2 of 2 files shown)


alt-glitch:

Duplicate of #54978: it already implements the same idle-session liveness gate for server-side `/steer` dispatch, with broader coverage of startup and rejected-steer cases.

liuhao1024:
Independently verified both PRs before responding.

**Where the triage is right:** #54978's behavior matrix does cover this PR's core fix — gating live injection on `session["running"]` is literally its first behavior — and it goes further on the startup window (running-but-agent-still-building) and rejected/failed steers via `queued_prompt`, which this PR deliberately leaves out (the residual race is acknowledged in the body).

**Why I'm keeping this PR open:** #54978's diff targets `tui_gateway/server.py:12162`, but slash-command dispatch has since been moved out of `server.py` into `tui_gateway/methods_tools.py` — the `/steer` branch now lives at `methods_tools.py:797` and current `server.py` no longer contains a `name == "steer"` dispatch branch. #54978's head predates that move (last push 2026-07-15, no rebase since), so it needs a full rewrite to land on current main. This PR applies the minimal idle-session gate at the current location and its checks are green.

**No objection either way:** if maintainers prefer #54978's broader approach once it is rebased onto `methods_tools.py`, closing this PR in its favor is fine by me — the startup-window handling there is genuinely better and would subsume this minimal gate.

PR #84927, opened 2026-08-13
feat(plugins): additive-only redaction pattern registry

teknium1:
## Summary
Plugins can now ship vendor token formats as redaction patterns via `ctx.register_redaction_patterns()` — additive-only by construction, so a plugin can only ever over-redact, never expose (sub-issue #65449, tracker #64182). Salvages PR #58550 (@webdevtodayjason) with authorship preserved.

## Changes
- Registry API: patterns must compile, must start with ≥2 literal characters (keeps the pre-screen substring gate effective; makes `.*`-style redact-everything patterns structurally impossible); no remove/unregister/weaken API for built-ins
- Salvage hardening on top: top-level alternation rejected (`ab|.*` can't smuggle a wildcard), ReDoS-shaped patterns rejected at registration (`(a+)+b` etc.), per-source pattern attribution + explicit pre-screen rebuild
- Reference plugin moved out-of-tree per the June 2026 no-third-party-plugins policy — the e2e test writes a synthetic plugin at test time instead of shipping `nvapi-redaction` in-tree
- Docs: plugins.md registry section with the over-redact-only contract

## Validation
| | Result |
|---|---|
| Tests | 195 passed (19 registry + 46 redaction-adjacent tool + 149 agent redaction/plugin API suites) |
| Invariants | `.*`, `(a)*`, `a[bc]+`, `ab\|.*`, `(a+)+b`, `[abc]{2,}` all REJECTED; built-ins unchanged after registration; live nvapi-style token actually masked |
| Lint | ruff clean |

Ground rule 4 satisfied: over-redaction is the only reachable failure mode.

## Infographic

redaction pattern registry

Issue #77465, opened 2026-08-03
fix(secrets): provider-egress redaction gaps — plain-.env secrets unmasked, encoding transforms bypass, uncovered assembly sites, post-mask mutation


andrexibiza:
## Severity: HIGH

**Cluster N-EGRESS** (7 findings against the shipped #77198 fix; Agent 3-restart + Agent 10).

The exact-value masker (`agent/redact.py:746 redact_known_secret_values`, wired at `tool_dispatch_helpers.py:735-744`) has gaps:

1. **Plain `.env` secrets under non-suffixed names are never collected** — the snapshot holds external-source values only (env_loader.py:161-166); `DATABASE_URL` in `.env` leaks in the default single-profile deployment. (HIGH — the masker's own headline case.)
2. **Multiplexed-gateway/cron profile `.env` secrets live in the ContextVar scope; the pass reads only os.environ** (redact.py:735).
3. **Encoding transforms defeat byte-exact matching**: base64/hex/URL/JSON-\u/case/whitespace via terminal, code_execution_tool, web_extract.
4. **Tool messages built WITHOUT make_tool_result_message bypass**: `agent/conversation_loop.py:7149` (str(e) exception → tool msg), `agent/mcp_tool.py:1354` (sampling results).
5. **Multimodal list items with type != "text" pass unmasked** (tool_dispatch_helpers.py:739-751); dict branch skips tuple/bytes.
6. **Post-mask mutation**: `/steer` appended after masking (agent_runtime_helpers.py:3888).
7. **Bare env names** (`TOKEN=`, `API_KEY=`) miss the underscore-required suffix list.

## Proposed fix
- Collect plain-`.env` values into the exact-value set (per-home), not just external-source values.
- Read the ContextVar-scoped home for the pass.
- Transform detection (base64/hex/URL) for the highest-risk tools.
- Route ALL tool-message assembly through the masker; cover `conversation_loop` and `mcp_tool`.
- Mask non-text parts; mask after `/steer` mutation; bare-name suffix matching.
- Gate tests for each.

## Related
#77198 (shipped fix — the base this extends).

andrexibiza:
**Overlap coordination (FILE-LIST found two open PRs in this issue's target surface):**
- #88606 (`fix(redact): add opt-in sensitive-data controls`) touches `agent/redact.py` — adding `_redact_extra_sensitive_key_values` opt-in controls (adjacent to, not identical with, finding #1's `redact_known_secret_values` masker).
[...]

Issue #77165, opened 2026-08-03
fix(security): applied-secrets snapshot not wired into provider-egress redaction (tool results, sanitized context, terminal output)


andrexibiza:
## Scope

The authoritative per-home applied-secrets snapshot — `_SECRET_SOURCE_VALUES_BY_HOME` / `get_secret_source_values()` (`hermes_cli/env_loader.py:42, 166`) — is populated by `_apply_external_secret_sources` but is not wired into the two remaining egress surfaces that carry secret values: **provider-bound tool-result content** (the wire) and **provider-bound sanitized context**. Both currently rely on shape-based redaction only.

## Defect

`redact_sensitive_text()` (`agent/redact.py:659`) masks by shape (vendor prefixes, URL credentials, auth headers, env-dump patterns). A tool result or context string containing an applied secret **value** with no recognizable shape — arbitrary 1Password item values, `DATABASE_URL`-style names, opaque `MY_SERVICE_TOKEN=abc123randomstring` — passes through unmasked and is transmitted to the model provider in the next API call.

Emission-side channels (status lines #77012, logs #77020) already use exact-value passes; the egress path does not. The exact-value primitive (`mask_known_secret_values`, `_CREDENTIAL_VALUE_SUFFIXES`) exists in `agent/redact.py` from #77020 but is only wired into `RedactingFormatter`.

## Affected egress points

- `agent/tool_dispatch_helpers.py:532` — `make_tool_result_message` (content assembled here, appended at `agent/tool_executor.py:650, 1234, 1327, 1356, 1927, 1994`)
- `agent/context_engine.py:42` — `sanitize_memory_context` (provider egress boundary)
- `agent/chat_completion_helpers.py:1459` — pre-send sanitized content
- `tools/terminal_tool.py:3047`, `tools/process_registry.py:2453` — `redact_terminal_output` (shape-based, no exact-value pass)

## Proposed fix

1. Add an exact-value pass to the tool-result egress path: mask `get_secret_source_values(home)` values + credential-suffixed env values in `make_tool_result_message` content before it enters `messages`.
2. Extend `redact_sensitive_text` (or add a wrapper) with the same exact-value pass for the pre-send sanitizers (`context_engine.py:42`, `chat_completion_helpers.py:1459`).
3. Extend `redact_terminal_output` to consult the applied-secrets snapshot, not just shapes.
[...]

Issue #77162, opened 2026-08-03
fix(security): exact-value applied-secret redaction missing on tool-result -> provider egress path


andrexibiza:
## Scope

Tool-result content assembled at `agent/tool_dispatch_helpers.py:532` (`make_tool_result_message`) and appended to the provider-bound `messages` list at `agent/tool_executor.py:650, 1234, 1327, 1356, 1927, 1994` is transmitted to the model provider API with no exact-value secret redaction.

## Current state

- Pre-send sanitization: `redact_sensitive_text()` (`agent/redact.py:659`) invoked at `agent/context_engine.py:42`, `agent/chat_completion_helpers.py:1459`, `agent/moa_loop.py:69`. Shape-based only: vendor prefixes, URL/query/userinfo credentials, auth headers, env-dump patterns. Opaque values lacking a recognizable shape pass through.
- Terminal path: `redact_terminal_output()` (`agent/redact.py:918`) at `tools/terminal_tool.py:3047`, `tools/process_registry.py:2453`. Same shape limitation.
- `mask_known_secret_values()` (merged #77020) provides exact-value masking for credential-suffixed env vars (`*_API_KEY/_TOKEN/_SECRET/_KEY/_PASSWORD`) but is wired only into `RedactingFormatter` (log egress).
- `get_secret_source_values()` / `_SECRET_SOURCE_VALUES_BY_HOME` (`hermes_cli/env_loader.py:42, 166`) holds the authoritative per-home snapshot of values applied from external secret sources (Bitwarden / 1Password / command — `agent/secret_sources/registry.py:167-183`). Never consulted at egress.

## Defect

A value applied from any external secret source, under any name — including non-credential-shaped names such as `DATABASE_URL`, `FOO`, arbitrary 1Password item keys — that a tool echoes into its result (config/env artifact read, `printenv`, backend error quoting a key) passes all shape-based passes and is transmitted verbatim in the provider request. Channel is provider-agnostic; applies to every registered secret source.

## Proposed fix

1. Exact-value pass on the tool-result egress path: mask `get_secret_source_values(home)` values + `_known_secret_values()` (credential-suffixed env) in `make_tool_result_message` content (`agent/tool_dispatch_helpers.py:532`) before `messages.append`.
2. Extend `redact_sensitive_text()` with the same exact-value pass for the pre-send sanitizers (`agent/context_engine.py:42`, `agent/chat_completion_helpers.py:1459`).
[...]

PR #97383, opened 2026-08-28
feat(redaction): registry-fed exact-value and key-form redaction from user pattern file

hermes-node8:
## What does this PR do?

**What this adds:** exact operator-registered values and registered key-forms are masked at Hermes' existing redaction boundaries, from a user-owned JSON pattern file.

**Why:** Hermes' built-in applied-secret snapshot covers external-source values only; operators cannot register their own exact secrets (values that arrive through non-`KEY=value` shapes, or short secrets exact matching can't safely touch).

**Safety:** additive-only (no pattern file → the pass is a no-op; under-masking is structurally impossible) · fail-safe (broken file → last-good set, never an unmasked gap) · sentinel-protected (masked values can never be written back) · display-only (never modifies `.env`, vault, or config) · per-instance (`$HERMES_HOME`-resolved default).

**Scope:** this PR adds the source/value layer; it does not modify at-rest data and does not implement egress-boundary wiring (#72778 is a deliberately separate follow-up — the two compose rather than conflict).

## Related Issue

- `Refs #43666` — addresses the **compaction-boundary slice** of the persistence-boundary redaction gap (sibling PR #43940 covers the DB-URI / mutable-reasoning slice). The issue has multiple open slices, so this is intentionally not a full "Fixes".
- Motivation: #77162 / #77165 / #77465 — exact-value applied-secrets gaps the internal snapshot cannot cover.

## Type of Change

- 🔒 Security fix

## Changes Made

Minimal integration surface — 5 code files + 2 test files:

- `agent/redact.py` — one new processing stage: `_redact_registry_patterns` (pattern-file load, mtime-keyed cache, fail-safe last-good, sentinel on file reads, longest-match-first literal pass, plain + URL-query key-form passes). Implementation is concentrated in the existing redaction module: no new subsystem, no new abstraction hierarchy, no new CLI, no new format persisted by Hermes; existing redaction semantics are unchanged when the pass is disabled.
- `cli.py` / `gateway/run.py` / `hermes_cli/main.py` — four small wiring changes: bridge `security.redact_patterns` (config.yaml fallback; the env var still wins) to `HERMES_REDACT_PATTERNS` at the three process entry points, before `agent.redact` is imported — mirroring the existing `security.redact_secrets` bridge.
- `hermes_cli/config.py` — documents the new key in the commented config template.
- `tests/agent/test_redact_registry_patterns.py` + `tests/hermes_cli/test_redact_patterns_bridge.py` — 26 tests: mask styles, fail-safe/cache/sentinel behavior, key-form forms, URL-boundary semantics (query-position key-form masking gated on `redact_url_credentials`, bounded at the next `&`/`#`; registered exact values unconditional), and the config→env bridge end-to-end.

### Security surface

No new network, IPC, subprocess, or write-side surface is introduced. The new security-relevant input is the operator-controlled pattern file — inside the agent's existing trust envelope (the same class as `config.yaml` / `.env`), processed fail-safe and additive-only. The generated matching expressions are ReDoS-resistant by construction: registered key names are escaped literals. Arbitrary user-supplied regex is **not** accepted by this PR.

### Why not a plugin?

Hermes' plugin redaction registry (#84927) lets plugin authors register token *formats*. The guarantees this feature requires — pass ordering relative to the built-in masker, URL-boundary semantics at engine call sites, unconditional execution at every boundary, operator-managed exact-value state — are not expressible through the current plugin contract (registration, not control).

## How to Test

1. Create a pattern file, e.g. `{"literals": ["my-exact-value"], "key_patterns": {"PIN": true}}`.
2. Point `HERMES_REDACT_PATTERNS` at it (or set `security.redact_patterns` in config.yaml).
3. Run the redaction-surface suite:
   `pytest tests/agent/test_redact.py tests/agent/test_compaction_redaction_boundaries.py tests/agent/test_tool_call_arg_no_redaction.py tests/hermes_cli/test_redact_config_bridge.py tests/agent/test_redact_registry_patterns.py tests/hermes_cli/test_redact_patterns_bridge.py tests/tools/test_terminal_output_transform_hook.py`
   (fork PRs: CI needs a maintainer's workflow approval to run — until then, exact repro:)
   `env -u HERMES_HOME HERMES_REDACT_PATTERNS=<pattern-file> python -m pytest tests/agent/test_redact.py tests/agent/test_compaction_redaction_boundaries.py tests/agent/test_tool_call_arg_no_redaction.py tests/hermes_cli/test_redact_config_bridge.py tests/agent/test_redact_registry_patterns.py tests/hermes_cli/test_redact_patterns_bridge.py tests/tools/test_terminal_output_transform_hook.py -q`

## Validation

- Rebased and verified on **current main (`18a76be1`, 2026-09-01)**: redaction-surface suite **142/142** with a populated pattern file, including the upstream terminal-output transform hook (`test_terminal_output_transform_still_runs_strip_and_redact`) and the PR's registry-pass + config-bridge tests.
- The registry key-form/query sub-passes respect the upstream `code_file` contract (#33801): gated on `not code_file or file_read`, so terminal/source-code output keeps its token-prefix markers; exact-value literals stay unconditional.
- Apply-checked across **v0.20.0 – v0.21.0 (8 release tags v2026.8.3 → v2026.8.31) + origin/main — APPLY OK** on all.
- Behavior battery (extremes, hermetic + populated pattern file): v0.20.0 **90/90**, v0.20.6 **114/114** (the transform failure is fixed on the shipped range too), v0.21.0 **116/116**.
- At-rest posture: the pattern file defaults to `$HERMES_HOME/state/redaction/redact_patterns.json` (0600, per-instance); file-read content is masked with a non-reusable sentinel (#35519 semantics), so the pattern file itself can never be written back over its own values.
- Compatibility is continuously checked against Hermes releases; the contributor re-verifies and rebases within 48 hours of new releases (latest check: 2026-09-01, v0.21.0).

## Checklist

- [x] Conventional commits
- [x] Tests run and green
- [x] Security surface documented
- [x] Cross-platform impact considered (pure Python, no new dependencies)


hermes-node8:
@kshitijk4poor — you've been working directly in agent/redact.py, so I'd especially value your take on whether this fits the current redaction direction. I've kept the integration surface narrow and aligned this revision with the newer URL-credential behavior.

hermes-node8:
Updated implementation: the registry pass now respects the `code_file` contract (#33801). With a populated pattern file, the previous revision re-masked KEY=value assignments in terminal/source-code output, destroying the token-prefix marker upstream guarantees survives — `test_terminal_output_transform_still_runs_strip_and_redact` failed deterministically. The key-form and URL-query sub-passes are now gated on `not code_file or file_read`; exact-value literals remain unconditional. Rebased on current main (`18a76be1`): the redaction-surface suite is **142/142** with a populated pattern file, and the apply matrix is green across v0.20.0 → v0.21.0 (8 tags) + origin/main. PR body updated with the exact repro (fork CI needs a maintainer's workflow approval to run).

hermes-node8:
Closing this PR — we found we can accomplish what we need through the plugin redaction registry (#84927): exact values register as escaped patterns with the built-in word-boundary anchors and file-read sentinel, so no engine change is required. Thanks for the consideration.

PR #97386, opened 2026-08-28
fix(cli): suppress restore prompt with --keep-stash

jbdrak:
## What does this PR do?

Suppresses the autostash restore prompt when `hermes update --keep-stash` is active.

## Related Issue

N/A. This is a focused UX correction for the existing `--keep-stash` update path.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes incorrect behavior)

## Changes Made

- Adds a shared prompt-decision helper for autostash restoration.
- Prevents `--keep-stash` from displaying a restore prompt that the update flow will ignore.
- Preserves the existing interactive prompt for normal updates.
- Adds regression tests for both behaviors.

## Why

`--keep-stash` intentionally leaves local changes parked in Git after a successful update. Before this change, Hermes still displayed `Restore local changes now? [Y/n]`, accepted the user's answer, and then skipped restoration anyway. That made the update output misleading and made a confirmed `Y` appear to be ignored.

## How to Test

- `python3 -m py_compile hermes_cli/update_cmd.py hermes_cli/main.py tests/hermes_cli/test_update_autostash.py`
- `uv run pytest -q tests/hermes_cli/test_update_autostash.py::test_update_keep_stash_does_not_prompt_to_restore tests/hermes_cli/test_update_autostash.py::test_update_restore_prompt_still_appears_for_interactive_restore`

Result: 2 passed.

The broader autostash test file was also exercised, but two existing full-update cases reached an unrelated local gateway-supervision failure (`ai.hermes.gateway` was not launchd-supervised in the test harness).


Enough1122:

**Overall:** Small refactor extracting stash-restore prompt guard for testability and correctly honoring `--keep-stash`.

**What it does**
- `hermes_cli/update_cmd.py` adds `def _should_prompt_for_stash_restore(auto_stash_ref, keep_stash, assume_yes, gateway_mode, stdin_isatty, stdout_isatty) -> bool` returning `auto_stash_ref is not None and not keep_stash and not assume_yes and (gateway_mode or (stdin_isatty and stdout_isatty))`.
- `_cmd_update_impl` now calls `_m()._should_prompt_for_stash_restore(...)` instead of inline `auto_stash_ref is not None and not assume_yes and (gateway_mode or (sys.stdin.isatty() and sys.stdout.isatty()))` — adds missing `not keep_stash` check so `--keep-stash` suppresses restore prompt (previously still prompted).
- `hermes_cli/main.py` re-exports `_should_prompt_for_stash_restore` for test import; `tests/hermes_cli/test_update_autostash.py` added/updated.

**Non-blocking notes**
- Return type `bool(...)` ensures `None` → `False` — correct. Extracted function enables unit test without TTY mocking.

No blocking issues.

jbdrak:
Rebased onto current main. Upstream had refactored this path into `_prepare_checkout_for_update` + `_CheckoutPlan`, so `prompt_for_restore` moved. I kept the extracted helper as-is and plumbed `keep_stash` through to that function, so the plan's single `prompt_for_restore` is now gated on `--keep-stash`.

Verified with `scripts/run_tests.sh` over the seven update test files: 142 passed / 0 failed, including the two new cases. Dropping the `keep_stash` term from the helper fails `test_update_keep_stash_does_not_prompt_to_restore`, so the tests do capture the behaviour change.

PR #97387, opened 2026-08-28
fix(stream): ignore malformed content and reasoning deltas

### LINKED ISSUE #97382: [Bug] AttributeError: 'int' object has no attribute 'lstrip' in _provider_stream_text_may_be_sse during streaming
### Bug Description

During chat completion streaming with certain OpenAI-compatible providers or upstream gateways that yield non-string data in stream chunks (e.g. integer tokens or token IDs in `delta.content`), Hermes Agent crashes with:

```text
WARNING agent.chat_completion_helpers: Streaming failed after partial delivery, not retrying: 'int' object has no attribute 'lstrip'
WARNING agent.chat_completion_helpers: Partial stream delivered before error; returning length-truncated stub with 0 chars of recovered content so the loop can continue from where the stream died: 'int' object has no attribute 'lstrip'
```

Because of this uncaught exception, Hermes interprets the broken stream as a token-limit length truncation, triggers 4 consecutive continuation attempts, and ultimately terminates with:
`⚠️ Processing stopped: Response remained truncated after 4 continuation attempts. Try again.`

---

### Root Cause

In `agent/chat_completion_helpers.py`:

```python
def _provider_stream_text_may_be_sse(text: str) -> bool:
    """Return True while pending text still looks like an SSE control block."""
    stripped = (text or "").lstrip()
```

If `text` is not a `str` (for example, `int`), `(text or "").lstrip()` throws `AttributeError: 'int' object has no attribute 'lstrip'`.

---

### Proposed Fix

Add defensive type checking to `_provider_stream_text_may_be_sse`:

```diff
-def _provider_stream_text_may_be_sse(text: str) -> bool:
+def _provider_stream_text_may_be_sse(text: Any) -> bool:
     """Return True while pending text still looks like an SSE control block."""
+    if not isinstance(text, str):
+        return False
     stripped = text.lstrip()
     if not stripped:
         return False
```

If `text` is non-string, it cannot be an SSE control block (`:keepalive`, `data:`, etc.), so returning `False` allows streaming to proceed normally without breaking execution.

1052326311:
## What does this PR do?

Fixes streaming corruption when an OpenAI-compatible provider sends a scalar non-string `delta.content`, `delta.reasoning_content`, or compatibility `delta.reasoning` value.

The shared chat-completions streaming boundary now rejects scalar non-text values before they enter text buffers, callbacks, or final joins. It deliberately does not stringify malformed integer token IDs into visible output. The current upstream list-content normalization is retained, so valid list-shaped text blocks continue to reach callbacks and the synthetic completion message.

## Related Issue

Fixes #97382

## Type of Change

- [x] Tests (adding or improving test coverage)

## Changes Made

- `agent/chat_completion_helpers.py`: preserve current-main list-content flattening while rejecting scalar non-text content, and validate both structured reasoning carriers at intake.
- `tests/run_agent/test_streaming.py`: cover malformed integer content plus both reasoning carriers followed by valid reasoning and text.

## How to Test

1. On current base `4209d371aa1bb8840ce8447555bdd863a1a96c38`, the real stream accumulator returns `42still delivered` for an integer content delta followed by a valid text delta: upstream normalization would otherwise stringify the malformed token ID.
2. On current head `1e6eca864ecdb6ee1d533cbceef73f837c72d806`, the same real path returns only `still delivered`; valid list-shaped content remains handled by the current-main regression.
3. Run `uv run --extra dev --extra anthropic --locked python -m pytest -q tests/run_agent/test_streaming.py` (38 passed, 4 skipped).
4. Run `uv run --extra dev --extra anthropic --locked ruff check agent/chat_completion_helpers.py tests/run_agent/test_streaming.py` and `uv run --extra dev --extra anthropic --locked python -m compileall -q agent/chat_completion_helpers.py tests/run_agent/test_streaming.py`.
5. Run `git diff origin/main...HEAD --check` and `trufflehog filesystem --no-update --only-verified agent/chat_completion_helpers.py tests/run_agent/test_streaming.py` (0 verified secrets).

### Code

- [x] I have read the Contributing Guide.
- [x] My commit messages follow Conventional Commits.
- [x] I searched existing open and merged PRs for the issue, exception text, owner, and stream-content terms.
- [x] This PR contains only the streaming-boundary fix and its regressions.
- [ ] I have run `pytest tests/ -q` and all tests pass. The focused streaming suite passed; the full repository suite was not run.
- [x] I have added regression tests.
- [x] I tested on macOS with Python 3.11.
- [x] Rebased on current `main` to resolve the current-head conflict without discarding the newly merged list-content behavior.

### Documentation and Housekeeping

- [x] Documentation update is N/A; the defensive behavior is internal.
- [x] `cli-config.yaml.example` update is N/A; no config changed.
- [x] `CONTRIBUTING.md` and `AGENTS.md` update is N/A; architecture and workflow did not change.
- [x] Cross-platform impact considered; the guard is pure Python wire-data validation.
- [x] Tool descriptions and schemas are N/A; tool behavior did not change.

## Screenshots / Logs

N/A. The exact base/candidate stream accumulator proof above shows malformed scalar values are no longer exposed while following valid content is delivered.

### DIFF
--- a/agent/chat_completion_helpers.py
+++ b/agent/chat_completion_helpers.py
@@ -4192,7 +4192,11 @@ def interruptible_streaming_api_call(agent, api_kwargs: dict, *, on_first_delta=
                 model_name = chunk.model
 
             # Accumulate reasoning content
-            reasoning_text = getattr(delta, "reasoning_content", None) or getattr(delta, "reasoning", None)
+            reasoning_text = getattr(delta, "reasoning_content", None)
+            if not isinstance(reasoning_text, str) or not reasoning_text:
+                reasoning_text = getattr(delta, "reasoning", None)
+            if not isinstance(reasoning_text, str):
+                reasoning_text = None
             if reasoning_text:
                 # Summary-part models (gpt-5.x and other Responses relays) send
                 # one complete markdown block per delta with no separator, so
--- a/agent/chat_completion_helpers.py
+++ b/agent/chat_completion_helpers.py
@@ -4209,8 +4209,14 @@ def interruptible_streaming_api_call(agent, api_kwargs: dict, *, on_first_delta=
             # Accumulate text content — fire callback only when no tool calls.
             # Some OpenAI-compatible providers emit a text delta as a list of
             # content blocks.  Convert it once so callbacks and the synthetic
-            # completion message always receive plain text.
-            delta_content = flatten_message_text(getattr(delta, "content", None), sep="")
+            # completion message always receive plain text. Ignore scalar
+            # non-text values rather than exposing malformed token IDs.
+            raw_delta_content = getattr(delta, "content", None)
+            delta_content = (
+                flatten_message_text(raw_delta_content, sep="")
+                if isinstance(raw_delta_content, (str, list))
+                else None
+            )
             if delta_content:
                 content_parts.append(delta_content)
                 if not tool_calls_acc:
--- a/tests/run_agent/test_streaming.py
+++ b/tests/run_agent/test_streaming.py
@@ -15,14 +15,14 @@ import pytest
 
 def _make_stream_chunk(
     content=None, tool_calls=None, finish_reason=None,
-    model=None, reasoning_content=None, usage=None,
+    model=None, reasoning_content=None, reasoning=None, usage=None,
 ):
     """Build a mock streaming chunk matching OpenAI's ChatCompletionChunk shape."""
     delta = SimpleNamespace(
         content=content,
         tool_calls=tool_calls,
         reasoning_content=reasoning_content,
-        reasoning=None,
+        reasoning=reasoning,
     )
     choice = SimpleNamespace(
         index=0,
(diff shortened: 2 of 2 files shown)


1052326311:
Current-head review request for `89b31097f57cd84ef3395553e26d56aae3c08aac`.

The PR is `MERGEABLE` against current base `93de1d3430a1cb955ef85715cf1f59581295fa21`; the two base commits added since the original reproduction do not touch `agent/chat_completion_helpers.py` or `tests/run_agent/test_streaming.py`.

On baseline `7d1c9aeab739842fa42112c95d086a9897f2d513`, an integer `delta.content` reaches `_provider_stream_text_may_be_sse()` and fails with `AttributeError: int object has no attribute lstrip`. The current head filters non-string values at the shared streaming boundary, so the following valid text chunk is still accumulated and emitted.

Exact-head validation passed:
- `scripts/run_tests.sh tests/run_agent/test_streaming.py` (35 passed, 4 skipped)
- `uv run ruff check agent/chat_completion_helpers.py tests/run_agent/test_streaming.py`
- `uv run python -m compileall -q agent/chat_completion_helpers.py tests/run_agent/test_streaming.py`
- `git diff HEAD^ HEAD --check` and a commit-range TruffleHog scan (0 verified secrets).

1052326311:
Updated current head to `ff8d76669ed0496219b643a5be3e7b45ba42771c` after the linked-issue follow-up exposed the same malformed-value failure in reasoning accumulation.

The prior head reproduced `TypeError: sequence item 0: expected str instance, int found` for both `delta.reasoning_content=42` and `delta.reasoning=42`. The follow-up commit filters non-string values at reasoning intake, preserves valid carrier fallback, and verifies that later reasoning and text deltas still reach the response and callbacks. It deliberately does not stringify malformed token IDs into user-visible text.

Exact-head validation passed:
- `python -m pytest -q tests/run_agent/test_streaming.py` (41 passed)
- `python -m ruff check agent/chat_completion_helpers.py tests/run_agent/test_streaming.py`
- `python -m compileall -q agent/chat_completion_helpers.py tests/run_agent/test_streaming.py`
- `git diff HEAD^ HEAD --check`
- TruffleHog filesystem scans of both changed files (0 verified secrets)

Requesting review of the current head.

1052326311:
Current-head review request for `1e6eca864ecdb6ee1d533cbceef73f837c72d806`.

This rebases #97387 onto current `main@4209d371aa1bb8840ce8447555bdd863a1a96c38` and resolves the conflict with the newly merged list-content normalization without losing either behavior: list-shaped text remains flattened, while scalar malformed token IDs are not stringified into visible output.

Real stream-accumulator proof on the exact base/head:
- base `4209d371...`: integer content then valid text returns `42still delivered`
- head `1e6eca...`: the same request returns `still delivered`

Exact-head validation passed:
- `uv run --extra dev --extra anthropic --locked python -m pytest -q tests/run_agent/test_streaming.py` (`38 passed, 4 skipped`)
- Ruff and `compileall` for both changed files
- `git diff origin/main...HEAD --check`
- TruffleHog changed-file scan (`0 verified secrets`)

Requesting review of the current head.

1052326311:
Closing as superseded by the current-main resolution confirmed by teknium1 in https://github.com/NousResearch/hermes-agent/issues/97382#issuecomment-5744101180. The content and reasoning normalization paths in 23bae43cfad and 37fb7adfd6d cover the reported crash. I am preserving this branch; changing the upstream coerce-versus-ignore choice would need the separate real SSE sample requested in that comment.

Issue #84437, opened 2026-08-12
Add a native Windows test-runner path for contributors


v1shay:
The repository documents native Windows support, but the canonical contributor test command is currently POSIX-only.

`scripts/run_tests.sh` assumes:
- Bash and `${BASH_SOURCE[0]}`
- a POSIX virtualenv at `.venv/bin` or `venv/bin`
- `env -i`
- POSIX locale values such as `C.UTF-8`

That means a native Windows contributor cannot run the documented test suite without Git Bash or a separate POSIX environment. A small `scripts/run_tests.ps1` wrapper, or a cross-platform Python entry point shared by both launchers, would make the Windows contributor path match the CI behavior while preserving the existing shell script for Unix-like systems.

This is separate from WSL support: the goal is a first-class native PowerShell path.

BkashJEE:
PR #97393 doesn't close this, but it removes the hard blocker underneath it: on Windows `pytest tests/` currently cannot finish collection at all.

pytest treats a collection error as fatal to the whole run rather than to the offending file, so two POSIX-only imports take the entire suite down:

- `tests/tools/test_bot_turn_lock.py:12` — `import fcntl` at module scope
- `tests/hermes_cli/test_doctor_journal_modes.py` — two `@pytest.mark.skipif(os.geteuid() == 0, ...)` decorators; `skipif` conditions are evaluated at collection time, so the unguarded call raises `AttributeError`. A third decorator in that same file already guards this with `hasattr(os, "geteuid")` and carries a comment naming the exact failure mode — the other two were missed.

Measured on Windows 11 with identical flags either side:

```
before:  40010/40062 collected, 2 errors, Interrupted
after:   40046/40098 collected, 0 errors
```

Both runs pass `--ignore=tests/acp --ignore=tests/acp_adapter`. Those 11 files fail with `ModuleNotFoundError: No module named 'acp'` — there's no `acp` package in the tree (only `acp_adapter/`, and `acp` isn't in the `packages` list in `pyproject.toml`), so that looks platform-independent and I left it alone. Someone who can check on Linux should confirm.

Two things still stand between this and a genuinely usable Windows runner, in case it's useful for scoping this issue:

1. Fixing collection exposes two pre-existing failures in `test_doctor_journal_modes.py` that were never reachable on Windows before — one asserts the POSIX string `"No such file or directory"` against `"[WinError 2] ..."`, the other asserts a forward-slash path against backslashes. Both are assertion-level platform assumptions.
[...]

PR #94383, opened 2026-08-25
test(bot-mode): skip the turn-lock module on Windows instead of aborting collection

Hotragn:
## The problem

`tests/tools/test_bot_turn_lock.py` (added by `ac3f9a2dc4`, #93091) imports `fcntl` unconditionally at module level. `fcntl` does not exist on Windows, so the import raises during collection — and an uncaught collection error is not "skip this file", it **aborts the entire run**:

```
ERROR collecting tests/tools/test_bot_turn_lock.py
tests\tools\test_bot_turn_lock.py:12: in <module>
    import fcntl
E   ModuleNotFoundError: No module named 'fcntl'
!!!!!!! Interrupted: 1 error during collection !!!!!!!
no tests collected, 1 error in 0.74s
```

That takes out **all of `tests/tools/`** on Windows — 7732 collectable tests — not just this one module. A Windows contributor cannot run the tools suite at all without `--ignore`ing this file by hand.

## The fix

Skip before the import, matching `tests/cli/test_termios_drift_heal.py`:

```python
if sys.platform == "win32":  # pragma: no cover
    pytest.skip("bot turn lock contention is POSIX flock-only", allow_module_level=True)

import fcntl
```

The tree's other convention — `pytestmark = pytest.mark.skipif(...)`, used by `tests/state/test_fts_rebuild_admission.py` — **cannot** work here: `pytestmark` is evaluated after the module body has already been imported, so the `fcntl` import would still raise first. That module gets away with it because it never imports `fcntl` itself; it spawns subprocesses instead.

## No coverage is lost

`acquire_turn_lock()` in `tools/bot_relay.py` already degrades to a no-op contextmanager on Windows (`try: import fcntl / except ImportError: yield`), precisely because `fcntl` is unavailable. The cross-process flock contention this module asserts on cannot occur there, so these tests could never have passed on Windows — they could only ever abort the run.

## Before / after

```
before:  no tests collected, 1 error
after:   7732/7739 tests collected (7 deselected); this module reports 1 skipped
```

POSIX behaviour is unchanged — the guard only fires on `win32`, and `import fcntl` still happens exactly as before everywhere else.

## Platforms

Verified on Windows 11 / CPython 3.11.

PR #97393, opened 2026-08-28
fix(tests): POSIX-only imports at collection time abort the whole suite on Windows

BkashJEE:
## What does this PR do?

`pytest tests/` cannot complete collection on Windows. Two modules touch POSIX-only APIs at import time, and pytest treats a collection error as fatal to the **entire run**, not to the offending file:

```console
$ python -m pytest tests --collect-only -q
ERROR tests/tools/test_bot_turn_lock.py
ERROR tests/hermes_cli/test_doctor_journal_modes.py
!!!!!!!!!!!!!!!!!! Interrupted: 13 errors during collection !!!!!!!!!!!!!!!!!!!
```

So a Windows contributor cannot run the suite at all — not a subset, not one file, nothing. That makes "please add a test with your fix" a hard ask on this platform.

Two import-time guards fix it. Both match a pattern that already exists in the tree.

## Related Issue

Related to #84437 (native Windows test-runner path for contributors). Not a full fix for it — this removes the hard blocker, it does not add a runner path.

## Type of Change

## Changes Made

**`tests/tools/test_bot_turn_lock.py`**
`import fcntl` at module scope raises `ModuleNotFoundError` on Windows. These tests exercise real flock contention between separate fds, so skipping the module is the honest outcome rather than faking the lock. Now `fcntl = pytest.importorskip("fcntl")` — the idiom already used at `tests/tools/test_file_sync_back.py:13`.

**`tests/hermes_cli/test_doctor_journal_modes.py`**
Two `@pytest.mark.skipif(os.geteuid() == 0, ...)` decorators call a POSIX-only function. `skipif` conditions are evaluated at collection time, so on Windows this raises `AttributeError` and takes the module down before any test runs.

A third decorator in the same file already guards this and carries a comment naming exactly this failure mode:

```python
@pytest.mark.skipif(
    # os.geteuid is POSIX-only, and a skipif condition is evaluated at
    # collection time — calling it unguarded would raise AttributeError
    # and take the whole module down on Windows.
    hasattr(os, "geteuid") and os.geteuid() == 0,
    reason="root ignores file permissions",
)
```

The other two occurrences were missed. Both now match it.

## How to Test

On Windows, with identical flags either side:

```console
# before
40010/40062 tests collected (52 deselected), 2 errors in 51.10s
!!!! Interrupted: 2 errors during collection !!!!

# after
40046/40098 tests collected (52 deselected) in 177.23s
```

```bash
python -m pytest tests --collect-only -q --ignore=tests/acp --ignore=tests/acp_adapter
```

On POSIX nothing changes: `importorskip` imports normally, and `hasattr(os, "geteuid")` is `True` so the skip condition evaluates exactly as before.

### Two notes I'd rather flag than bury

**1. The `acp` collection errors are probably not a platform issue.** Both runs above pass `--ignore=tests/acp --ignore=tests/acp_adapter`, which fail with `ModuleNotFoundError: No module named 'acp'`. There is no `acp` package in the tree — only `acp_adapter/`, and `acp` is not in the `packages` list in `pyproject.toml`. That looks like it would fail everywhere, not just on Windows, so I have not touched it. Worth a look from someone who can confirm on Linux.

**2. Fixing collection exposes two pre-existing Windows failures** in `test_doctor_journal_modes.py` that had never been reachable on this platform:

- `test_missing_file_keeps_the_os_error_text` asserts the POSIX string `"No such file or directory"`; Windows produces `"[WinError 2] The system cannot find the file specified"`.
- `test_lists_every_managed_database` asserts `"kanban/boards/myboard/kanban.db"`; Windows produces backslash separators.

Both are assertion-level platform assumptions, a different class of problem from a collection blocker, and fixing them means choosing how strict those assertions should be. I've left them out rather than widening this PR's scope, and I'm happy to send a follow-up if you tell me which way you'd want them written.

### Code

- [x] I've run the relevant tests and they pass (`tests/tools/test_bot_turn_lock.py` skips cleanly on Windows; `test_doctor_journal_modes.py` is 31 passed / 4 skipped / 2 pre-existing failures documented above)
- [x] N/A — this fixes test collection itself; the verification is the collection run above

- [x] N/A — no user-facing behavior change
- [x] N/A — no config keys
- [x] Cross-platform impact considered: both changes are no-ops on POSIX by construction

### DIFF
--- a/tests/hermes_cli/test_doctor_journal_modes.py
+++ b/tests/hermes_cli/test_doctor_journal_modes.py
@@ -142,7 +142,13 @@ class TestReadJournalMode:
             holder.close()
 
     @pytest.mark.skipif(os.name == "nt", reason="chmod is a no-op on Windows")
-    @pytest.mark.skipif(os.geteuid() == 0, reason="root ignores file permissions")
+    @pytest.mark.skipif(
+        # os.geteuid is POSIX-only, and a skipif condition is evaluated at
+        # collection time — calling it unguarded would raise AttributeError
+        # and take the whole module down on Windows.
+        hasattr(os, "geteuid") and os.geteuid() == 0,
+        reason="root ignores file permissions",
+    )
     def test_read_only_directory_is_still_readable(self, tmp_path):
         db = tmp_path / "state.db"
         _make_db(db, journal_mode="WAL")
@@ -387,7 +393,13 @@ class TestReportDatabaseJournalModes:
         assert "state.db: rollback journal mode" in out
 
     @pytest.mark.skipif(os.name == "nt", reason="chmod is a no-op on Windows")
-    @pytest.mark.skipif(os.geteuid() == 0, reason="root ignores file permissions")
+    @pytest.mark.skipif(
+        # os.geteuid is POSIX-only, and a skipif condition is evaluated at
+        # collection time — calling it unguarded would raise AttributeError
+        # and take the whole module down on Windows.
+        hasattr(os, "geteuid") and os.geteuid() == 0,
+        reason="root ignores file permissions",
+    )
     def test_unreadable_database_does_not_crash(self, tmp_path, capsys):
         db = tmp_path / "state.db"
         _make_db(db)
--- a/tests/tools/test_bot_turn_lock.py
+++ b/tests/tools/test_bot_turn_lock.py
@@ -9,7 +9,6 @@ process, so threads exercise the true kernel-lock semantics.
 
 from __future__ import annotations
 
-import fcntl
 import json
 import os
 import re
@@ -17,6 +16,13 @@ import threading
 import time
 
 import pytest
+
+# fcntl is POSIX-only. Importing it at module scope raised
+# ModuleNotFoundError on Windows during collection, which pytest counts as
+# a collection error and which aborts the whole run — not just this file.
+# These tests exercise real flock contention, so skipping the module is the
+# honest outcome; same idiom as tests/tools/test_file_sync_back.py.
+fcntl = pytest.importorskip("fcntl")
 
 from tools import bot_mode_dm, bot_relay
 from tools.bot_relay import TurnBusyError, acquire_turn_lock, turn_lock_path


alt-glitch:

Related: #94383 already repairs the bot-turn-lock fcntl import and #95122 repairs the doctor os.geteuid collection guard. This PR combines both focused Windows test fixes; please consolidate with those open efforts.

BkashJEE:
Triage is right — this duplicates #94383 and #95122, which both predate it and are the better shape. Closing in favour of them.

For anyone landing here: those two are focused, one file each, CI green, and #95122 goes further than this PR did. I documented `test_missing_file_keeps_the_os_error_text`'s hardcoded `"No such file or directory"` as out of scope; #95122 actually fixes it, and does it properly by taking the expected string from `excinfo.value.strerror` rather than pinning either platform's wording. #94383's `pytest.skip(..., allow_module_level=True)` is also better reasoned than the `importorskip` here — it cites `tests/cli/test_termios_drift_heal.py` as precedent and notes that `acquire_turn_lock()` degrades to a no-op on Windows anyway, so there is genuinely nothing to run.

Two things from my measurements that may still be useful to whoever reviews those PRs:

Collection on Windows 11, identical flags either side of the fix:

```
before:  40010/40062 collected, 2 errors, Interrupted
after:   40046/40098 collected, 0 errors
```

Both runs pass `--ignore=tests/acp --ignore=tests/acp_adapter`; those 11 files fail with `ModuleNotFoundError: No module named 'acp'` and there is no `acp` package in the tree (only `acp_adapter/`, and `acp` is not in `pyproject.toml`'s `packages`), so that looks platform-independent and separate from either PR.

And once collection works, `scripts/check-windows-footguns.py tests` reports ~2.3k matches, of which roughly 55 are behavioural rather than cosmetic: 19 `os.killpg`, 12 `os.getuid`/`geteuid`, 10 `signal.SIGKILL`, 8 `os.kill(pid, 0)`, 6 `os.setsid`. Those are what would keep failing on Windows after #94383 and #95122 land.

Sorry for the noise — I should have searched for the fix terms, not just the symptom.

PR #97399, opened 2026-08-28
feat(gateway): add bounded Telegram profile-turn routing

thebrettwatson:
## Summary
- add opt-in, per-message Telegram profile-turn routing without multiplexing or secondary adapters
- make configured aliases self-waking only for leading group text after existing auth/chat/topic/guest/ownership gates
- visibly refuse embedded aliases, captions/media, and commands instead of falling through to the default profile
- visibly reject invalid, disabled, or unserved targets
- preserve legacy behavior when no route configuration exists
- isolate target profile config, session namespace, and credentials for the routed turn

## Verification
- dedicated profile-turn module — 17 passed
- all 58 Telegram modules — 587 passed
- 9 coupled config/profile/session modules — 153 passed
- Ruff check and format check — PASS
- `git diff --check origin/main...HEAD` — PASS

## Runtime boundary
No live config, gateway process, Telegram message, or fleet topology was changed. Merge/install/restart and first-fire receipts remain separate gates.

### DIFF
--- a/agent/secret_scope.py
+++ b/agent/secret_scope.py
@@ -56,6 +56,13 @@ def is_multiplex_active() -> bool:
 _SECRET_SCOPE: ContextVar[Optional[Mapping[str, str]]] = ContextVar(
     "_SECRET_SCOPE", default=None
 )
+# A profile turn on an otherwise single-profile gateway needs the same
+# credential isolation as a multiplexed turn, without changing the process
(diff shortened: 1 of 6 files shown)


thebrettwatson:
Closing this upstream proposal because the capability is operator-specific to a local multi-profile fleet. The verified implementation will remain maintained in the contributor fork rather than expanding the upstream core surface. Thank you.

PR #97402, opened 2026-08-28
fix(update): preserve gateway progress logging

### LINKED ISSUE #97394: Desktop hand-off watchdog cancels healthy Windows updates: `--gateway` mode never creates logs/update.log, the only progress signal windows.ps1 watches
## Summary

On Windows the Desktop **Update** button hands off to `scripts/desktop-update/windows.ps1`,
which runs `python -m hermes_cli.main update --yes --gateway ...` under a step-idle watchdog.
That watchdog treats **growth of `logs/update.log`** as the progress signal for steps that are
stdout-silent (the Electron/desktop rebuild).

But in `--gateway` mode `logs/update.log` is **never created at all**:
`hermes_cli/main.py::_install_hangup_protection` returns early for `gateway_mode` *before* the
update.log mirror is installed. The watchdog is therefore watching a file that cannot exist on
the exact code path the Desktop uses.

Consequence: any Desktop-driven update whose desktop rebuild exceeds
`$script:StepIdleTimeoutSeconds` (600 s default, overridable via
`HERMES_UPDATE_STEP_IDLE_SECONDS`) is cancelled as "stalled" with **exit 124**, even though the
build is healthy — and the build output is lost too, because `update_cmd.py::_log_only_write`
writes to `sys.stdout._log`, which is `None` in gateway mode (output discarded, not merely
unlogged).

## Evidence (real run, Windows 11)

```
logs/desktop-update-handoff.log
19:17:08  running: python -m hermes_cli.main update --yes --gateway --force --branch main --keep-stash
19:28:20  update!| step stalled: no stdout/stderr for 600s and no update.log growth while pid 29904
          remained alive; cancelling its process tree.
          hermes update exit code: 124
```

* `logs/update.log` mtime = **2026-08-15**, i.e. the last CLI-mode run. No Desktop-driven
  update has ever written that file on this machine.
* Nothing was actually blocked: the Electron build was progressing; only the witness was absent.
* Earlier hand-offs on this machine all exited 0 — they either reported `desktop skipped` or
  finished the rebuild in under 10 minutes. The first rebuild that ran longer than the idle
  ceiling was the first kill.

## Reproduction

1. Windows, installation with `apps/desktop` present.
2. Click Settings -> About -> Update (this is the `--gateway` hand-off path).
3. Make the desktop rebuild step exceed 600 s of stdout silence (slow disk, AV scanning
   `node_modules`, cold npm cache).
4. -> `step stalled ... no update.log growth`, `exit code: 124`, repo left mid-update.

## Code pointers

* `hermes_cli/main.py::_install_hangup_protection` — `if gateway_mode: return state` fires
  before the update.log mirror is installed.
* `scripts/desktop-update/windows.ps1` — `$script:StepProgressLogPath = Join-Path $LogDir "update.log"`;
  the idle clock resets only on growth of that path; `$script:StepIdleTimeoutSeconds = 600`.
* `update_cmd.py::_log_only_write` — targets `sys.stdout._log`, which is `None` in gateway mode.

## Proposed fix

Install the update.log mirror in **both** modes and keep only the SIGHUP handling
terminal-specific. The hangup behaviour is what is mode-specific; the log mirror is not:

```python
    if not gateway_mode:
        _install_sighup_ignore()

    return _install_update_log_mirror(state)
```

Alternatively (or additionally): have `windows.ps1` watch a signal that provably exists in
gateway mode — a heartbeat file written by `_run_logged_subprocess`, or the desktop release
directory mtime — so the watchdog cannot be defeated by a missing witness.

## Workaround (verified locally)

1. `HERMES_UPDATE_STEP_IDLE_SECONDS=3600` in the User environment.
2. The one-line change above; verified with `hermes update --check --gateway`, which then
   appends to `logs/update.log` as expected.

## Related, not duplicates

* #96409 — same watchdog, but the trigger there is a genuine npm stall plus the
  half-updated-repo / Repair-install recovery gap. This report is about the watchdog's progress
  witness being structurally unavailable in `--gateway` mode, so healthy builds get killed.
* #95897 / #95970 — introduced the watchdog and raised its default to 600 s. Raising the
  ceiling only delays this failure mode; the file being watched still never appears.

fangliquanflq:
## Summary

- keep `logs/update.log` mirroring active for `hermes update --gateway`
- retain SIGHUP handling only for terminal-driven updates
- cover gateway progress and log-only desktop build output with a regression test

## Root cause

`_install_hangup_protection()` returned before installing the output mirror in gateway mode. Windows Desktop's update hand-off watches `logs/update.log` growth during stdout-silent build steps, so the structurally missing log made a healthy long-running build appear stalled and allowed the watchdog to terminate it with exit 124. The same early return also left `_log_only_write()` without a log target, discarding captured desktop build output.

## Verification

- `scripts/run_tests.sh tests/hermes_cli/test_update_hangup_protection.py tests/hermes_cli/test_update_handoff_exit.py -q` (17 passed)
- `scripts/run_tests.sh tests/test_desktop_update_windows_pipe_drain.py tests/test_desktop_update_windows_python_handoff.py -q` (10 passed)

Closes #97394

### DIFF
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -10633,7 +10633,9 @@ def _install_hangup_protection(gateway_mode: bool = False):
     signals the user or OS sent on purpose.
 
     In gateway mode (``hermes update --gateway``) the update is already
-    spawned detached from a terminal, so this function is a no-op.
+    spawned detached from a terminal, so SIGHUP protection is unnecessary.
+    Output is still mirrored to ``update.log`` because the Windows Desktop
+    hand-off watchdog uses growth of that file as its progress signal.
 
     Returns a dict that ``cmd_update`` can pass to
     ``_finalize_update_output`` on exit.  Returning a dict rather than a
@@ -10646,19 +10648,17 @@ def _install_hangup_protection(gateway_mode: bool = False):
         "installed": False,
     }
 
-    if gateway_mode:
-        return state
+    if not gateway_mode:
+        import signal as _signal
 
-    import signal as _signal
-
-    # (1) Ignore SIGHUP for the remainder of this process.
-    if hasattr(_signal, "SIGHUP"):
-        try:
-            _signal.signal(_signal.SIGHUP, _signal.SIG_IGN)
-        except (ValueError, OSError):
-            # Called from a non-main thread — not fatal.  The update still
-            # runs, just without hangup protection.
-            pass
+        # (1) Ignore SIGHUP for the remainder of this process.
+        if hasattr(_signal, "SIGHUP"):
+            try:
+                _signal.signal(_signal.SIGHUP, _signal.SIG_IGN)
+            except (ValueError, OSError):
+                # Called from a non-main thread — not fatal.  The update still
+                # runs, just without hangup protection.
+                pass
 
     # (2) Mirror output to update.log and wrap stdio for broken-pipe
     # tolerance.  Any failure here is non-fatal; we just skip the wrap.
@@ -10799,9 +10799,8 @@ def cmd_update(args):
 
     gateway_mode = getattr(args, "gateway", False)
 
-    # Protect against mid-update terminal disconnects (SIGHUP) and tolerate
-    # writes to a closed stdout.  No-op in gateway mode.  See
-    # _install_hangup_protection for rationale.
+    # Protect terminal updates against SIGHUP and mirror both terminal and
+    # gateway update output to update.log. See _install_hangup_protection.
     _update_io_state = _install_hangup_protection(gateway_mode=gateway_mode)
     # Cross-process mutual exclusion. The dashboard's Update button spawns
     # this same command detached, and the desktop hands off to the Tauri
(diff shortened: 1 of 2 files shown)


Artemonim:
Layer 2 (stream `update.log` as the child runs, progress heartbeats during silent npm/Electron, gateway watcher 3600s) is https://github.com/NousResearch/hermes-agent/pull/101850. It leaves the `--gateway` tee to this PR and is written to merge after it.

fangliquanflq:
Thanks for mapping the dependency. I verified that #101850 leaves the gateway-mode tee owned by this PR and adds the separate live-child streaming, silent-step heartbeat, and 60-minute gateway-watcher layer. No change is needed in #97402: its current head only makes SIGHUP handling terminal-specific while keeping the `update.log` mirror active in gateway mode, and its regression test covers both gateway stdout and log-only desktop-build output. On this head, the Python test suite and Windows-only CI passed; the blocking lint wrapper failed before lint execution because `setup-uv` could not fetch its manifest.

PR #96687, opened 2026-08-27
fix: don't hard-exit hermes desktop when sudo has no TTY

smfworks:
## Bug Description

`hermes desktop` at Hyprland/uwsm autostart dies in ~3s. Interactive `sudo chown/chmod` on `chrome-sandbox` has no TTY (`conversation failed`), then `sys.exit(1)` because Arch/Omarchy has no `apparmor_restrict_unprivileged_userns` sysctl, so the `--no-sandbox` fallback never ran.

## Root Cause

`_desktop_linux_sandbox_fixup` called `sudo` without `-n`. On failure, `--no-sandbox` was gated on `_desktop_linux_needs_no_sandbox()`, which is False when that sysctl file is missing.

## Fix

- `sudo -n` for the sandbox helper
- If the helper file exists but could not be configured, always launch with `--no-sandbox` (do not require the AppArmor sysctl)

## How to Verify

1. Linux host without `/proc/sys/kernel/apparmor_restrict_unprivileged_userns`
2. `chrome-sandbox` not root:root 4755
3. `hermes desktop` with no TTY (`setsid hermes desktop </dev/null`) should print the fallback warning and launch, not exit 1

## Test Plan

- [x] Added regression tests in `tests/hermes_cli/test_gui_command.py` (2 passed)
- [ ] Existing tests still pass (CI)
- [x] Manual: Omarchy autostart of unpacked `Hermes --no-sandbox` survives reboot; this PR is the official launcher equivalent

## Risk Assessment

Low — only the Linux packaged-desktop launch path. Successful SUID helper setup is unchanged. `--no-sandbox` was already the documented fallback on AppArmor-restricted hosts.

PR #50947, opened 2026-06-22
fix(desktop): skip chrome-sandbox setup when no TTY is available

riceharvest:
When `hermes desktop` is launched from a background process, cron, autostart, or any non-interactive shell, `_desktop_linux_sandbox_fixup` tries `sudo chown root:root` on `chrome-sandbox`, which fails with `a terminal is required to read the password` and aborts the entire desktop launch.

Check `sys.stdin.isatty()` before attempting sudo. When no TTY is available, print a warning with the manual setup command and return True (skip) instead of False (exit). Also change the sudo-not-found and sudo-failed paths to return True — warning instead of hard-failing.

```
hermes_cli/main.py | 16 +++++++++++++---
1 file changed, 13 insertions(+), 3 deletions(-)
```

PR #97404, opened 2026-08-28
fix(desktop): fall back to --no-sandbox when launcher runs non-interactively

wdmason:
## Bug

`hermes desktop` hard-fails (`sys.exit(1)`) in non-interactive/headless contexts such as a systemd user service or cron because it cannot `sudo chown root:root` + `sudo chmod 4755` Electron's Linux `chrome-sandbox` setuid helper — there is no TTY to prompt for the sudo password, so the subprocess dies and the service crash-loops with `status=1/FAILURE`.

The launcher already has a graceful `--no-sandbox` fallback, but it was gated behind `_desktop_linux_needs_no_sandbox()`, which only returned `True` when `ELECTRON_DISABLE_SANDBOX=1` was set **or** the host had `/proc/sys/kernel/apparmor_restrict_unprivileged_userns == "1"`. On a plain Linux host with neither, the fallback never fired and the launcher exited 1.

## Fix

Extend `_desktop_linux_needs_no_sandbox()` to also return `True` when the process is running non-interactively (`sys.stdin` is not a TTY), because in that context there is no way to prompt for the sudo password to configure the setuid sandbox helper — so the caller falls back to `--no-sandbox` rather than hard-failing into a crash loop.

- New check is placed **after** the `sys.platform != "linux"` guard and the `euid == 0` guard (root/non-Linux behavior unchanged), and **independent of** the AppArmor file check.
- The existing `sys.exit(1)` hard-fail branch is preserved for genuinely broken cases (helper missing / not a regular file) and for interactive contexts where the user could fix it.
- Fallback warning message generalized to cover both non-interactive launches and AppArmor-restricted hosts.

## Tests

- Unit test `test_desktop_linux_needs_no_sandbox_when_stdin_not_a_tty`: monkeypatches `sys.stdin` to a non-TTY object, clears `ELECTRON_DISABLE_SANDBOX`, and asserts `_desktop_linux_needs_no_sandbox()` is `True` on Linux (`False` on darwin/win32).
- `cmd_gui`-level test `test_gui_falls_back_to_no_sandbox_when_helper_cannot_be_configured`: patches `_desktop_linux_sandbox_fixup` to return `False`, keeps a regular `chrome-sandbox` file in the packaged tree, and asserts the launch subprocess is invoked with `--no-sandbox` and does **not** `sys.exit(1)` before launching.

`pytest tests/hermes_cli/test_gui_command.py` → `43 passed, 9 skipped` (one pre-existing ozone-hint test is host-env-dependent and fails only when `ELECTRON_OZONE_PLATFORM_HINT` is exported; it fails identically on `main`).

### DIFF
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -8090,6 +8090,14 @@ def _desktop_linux_needs_no_sandbox() -> bool:
         return False
     if hasattr(os, "geteuid") and os.geteuid() == 0:
         return False
+
+    # Non-interactive launch (e.g. a systemd user service or cron): there is no
+    # TTY to prompt for the sudo password that configuring the setuid
+    # chrome-sandbox helper requires, so fall back to --no-sandbox rather than
+    # hard-failing into a restart crash-loop.
+    if sys.stdin is not None and not sys.stdin.isatty():
+        return True
+
     try:
         with open("/proc/sys/kernel/apparmor_restrict_unprivileged_userns", encoding="utf-8") as f:
             return f.read().strip() == "1"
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related to #96687 and #50947: this focuses on the no-TTY predicate, while #96687 provides the broader helper-failure fallback.

monerostar (commented):
Ubuntu 26.04 on a local 5800X box. This overnight job is non-interactive (sys.stdin.isatty() is False, euid 1000). Host /proc/sys/kernel/apparmor_restrict_unprivileged_userns is already 1, so main already returns True here.

Forced that AppArmor file to 0 and imported _desktop_linux_needs_no_sandbox from each tree with the same non-tty stdin:

- main (25fcc8ad): False, so a failed sandbox fixup would still sys.exit(1)
- this PR (e3e08114): True

pytest on the PR tree, addopts cleared:
  test_desktop_linux_needs_no_sandbox_when_stdin_not_a_tty
  test_gui_falls_back_to_no_sandbox_when_helper_cannot_be_configured
  2 passed in 0.75s

Looks good for the no-TTY fallback. Related #96687 covers the broader helper-failure path.

Issue #95294, opened 2026-08-26
[Bug]: interrupted 'hermes update' (pull done, restart never ran) leaves gateway on stale code forever — receipt records the skew but no later run repairs it


teknium1:
## Summary

An interrupted `hermes update` (Ctrl-C / SIGHUP-death) between "git pull succeeded" and "fleet restart ran" leaves running gateways on stale code indefinitely — and **nothing ever repairs or even mentions the skew afterward**, even though the update receipt on disk records it exactly. On Teknium's box this broke every LLM cron job for ~2 days (Aug 24 09:00 → Aug 25 21:15) with `ImportError: cannot import name 'user_originated_turn_view' from 'agent.context_compressor'` on every fire.

This is the **interrupt** member of the "gateway serves stale code after update" class (siblings: #92145 restart-phase ImportError abort, #85753 macOS autostash prompt stall, #86207 dashboard variant; umbrella #91277). The distinguishing gap here: the skew was *known and persisted* in `update_receipts/latest.json`, and no subsequent `hermes` run acted on it.

## Real-world incident timeline (from `~/.hermes/logs/update.log` + `update_receipts/`)

| When (Aug 2026) | Event |
|---|---|
| 23 23:50 | `hermes update` pulls 46 commits, **restarts gateway cleanly** → PID 2111768 boots on `7e67f64f` |
| 24 03:36 | `hermes update` pulls **108 commits** (includes `ebae0064a`, which adds `user_originated_turn_view` to `agent/context_compressor.py` + lazy call sites in `agent/turn_context.py`). Run gets through pycache sweep + lazy-backend refresh, then dies **before** "Code updated!" / fleet restart. No receipt, no error line — Ctrl-C or terminal death. |
| 24 13:35 | Another `hermes update`, Ctrl-C'd during "Fetching updates...". Its receipt (below) **captures the skew explicitly** and then nothing uses it. |
| 24 09:00 → 25 21:15 | Every agent cron job fails: gateway's cached `agent.context_compressor` (old) + freshly lazy-imported `agent/turn_context.py` (new, from disk) → `cannot import name 'user_originated_turn_view'`. 15 jobs missed. no_agent watchdog jobs unaffected. |
| 25 21:15 | Manual `hermes gateway restart` fixes it instantly. |

## The smoking-gun receipt

`~/.hermes/logs/update_receipts/update_20260824_133554_2903266.json` (also `latest.json`):

```json
{
  "plan": {
[...]

Issue #86207, opened 2026-08-14
[Bug]: after `hermes update`, systemd-supervised dashboard keeps running stale code → Models page `/api/model/options` 500 (ImportError), needs manual restart


jimmywu0816:
## Bug Description

After `hermes update` (v0.20.1, commit f52feed1e, 2026-08-14), the **dashboard** (systemd user unit `hermes-dashboard.service`, running `hermes dashboard --skip-build`) kept running the pre-update code. One day later, the Models page model picker (Set main model) started failing with:

```
500: {"detail":"Failed to list model options"}
```

Root cause: `hermes_cli/web_server.py` `/api/model/options` -> `hermes_cli/inventory.py` -> `hermes_cli/models.py::model_supports_fast_mode()` does a function-level `from agent.model_metadata import is_grok_46_family`. That name was added to `agent/model_metadata.py` **by the update**, but the dashboard process had imported `agent.model_metadata` at startup (before the update) and Python serves the stale module from `sys.modules` forever. The dashboard never gets restarted by the update flow (it is a separate unit from the gateways), so it crashes at that import on the first picker open.

The update flow restarts gateways but not the dashboard unit. On this host the gateways were all restarted as part of the update and are healthy; only the dashboard went stale.

## Steps to Reproduce

1. Have `hermes-dashboard.service` (systemd user unit) running across a `hermes update`.
2. Update lands new names in modules the dashboard imported at startup (e.g. `is_grok_46_family` in `agent/model_metadata.py`).
3. Open dashboard -> Models page -> model picker (or call `GET /api/model/options`).
4. Receive `500: {"detail":"Failed to list model options"}`.

## Expected Behavior

Either the update flow restarts the dashboard (like it restarts gateways), or the dashboard detects code skew and surfaces a clear "restart required" state instead of a 500 on a core page.

## Actual Behavior

Silent stale process; `/api/model/options` fails with a generic 500 until the dashboard is manually restarted (`systemctl --user restart hermes-dashboard`), after which everything works (verified: 200, all provider rows listed, including across `?profile=` scopes).

## Environment

- OS: Linux (Ubuntu-based), systemd user units for both gateway and dashboard
- Version/Commit: v0.20.1 / f52feed1e
- Python: 3.11.15
- Browser: n/a (API-level repro)

## Error Output

Dashboard log (`~/.hermes/logs/gui.log`):

```
GET /api/model/options failed
Traceback (most recent call last):
  ...
[...]

Issue #69754, opened 2026-07-23
[Bug]: Code-skew guard blocks model switching on every profile after an update, but offers no auto-resolve / no auto-restart option


asubektibdg22-source:
# [Bug]: Code-skew guard blocks model switching on every profile after an update, but offers no auto-resolve / no auto-restart option

**Repository:** NousResearch/hermes-agent
**Component:** gateway, cli (install-update)
**Type:** bug (UX / reliability gap)
**Severity:** P2 (impacts multi-profile / multi-gateway-token users on every update)

---

## Summary

After `hermes update` (or a manual `git pull`) updates the shared checkout, the
gateway's `_model_switch_skew_guard()` refuses `/model` switches on **every
profile** with:

> This gateway is running code from `<boot_rev>` but the checkout on disk is now
> `<disk_rev>`. Switching models would risk a stale-module crash — restart the
> gateway to load the new code: `hermes gateway restart`

The guard is correct and valuable (it replaced the old cryptic
`cannot import name 'env_float' from 'utils'` crash, see #51564). But the
**recovery path is fully manual**, and there is **no config option to
auto-restart the gateway after an update or on skew detection**. On a
multi-profile host this means every update leaves the user one manual step away
from being unable to switch models in *any* profile.

## Why it affects all profiles (not just one)

From `gateway/code_skew.py`:

- The gateway is a **single long-lived process** (`gateway run --replace`); its
  `sys.modules` is frozen at boot.
- `record_boot_fingerprint()` snapshots the checkout rev **once at startup**.
- `detect_code_skew()` compares that boot fingerprint against the **current
  checkout on disk** (`_PROJECT_ROOT`, a single shared directory used by all
  profiles).

So the comparison is global: if the disk checkout drifted since boot, the guard
fires for a `/model` switch in **any** profile routed through that gateway
(`default`, `hmi1`, `pinat`, `parklife22`, ...). The guard is intentionally
scoped only to `/model` switching (the highest-risk lazy-import trigger), so a
session that isn't switching models keeps working — but as soon as any profile
tries to change models it gets blocked.

## Additional friction on multi-profile / shared-token setups

When two profiles share the **same Telegram bot token** (only one gateway may
bind it), the documented recovery is:

```
hermes gateway stop --profile default
hermes gateway restart --profile pinat
```

`hermes update` does restart the gateway at the end, but it restarts only the
active/default gateway — it does **not** coordinate the sibling gateway that
[...]

PR #66999, opened 2026-07-18
fix(gateway): reload clean code after checkout drift

chromalinx:
## Summary

Reload a gateway safely when its checkout changes underneath a long-lived process.

## Root cause

A gateway retains imported modules for its lifetime. After a checkout update, a later lazy import can combine new source with an old in-memory dependency. Cached bytecode can also survive into the next boot when the checkout changed.

## Fix

- Persist the gateway boot fingerprint and remove stale gateway bytecode before supported launchers import `gateway.run`.
- Trigger a graceful reload for `/model` code skew instead of only refusing the command.
- Use the shared supervisor detector so launchd, systemd, s6, and externally supervised gateways select the service restart path.
- Reject direct `python -m gateway.run` execution because it cannot establish the required pre-import bootstrap boundary.

## Non-goals

This does not replay the original `/model` command after reload. The gateway reports that the reload is in progress and asks the user to re-run the command.

## Test plan

- `pytest tests/test_code_skew.py -q -o 'addopts='` — 23 passed
- `ruff check` on changed Python files
- `py_compile` on changed Python and launcher files
- `git diff --check`
- Independent CoVe review — PASS
- Local runtime activation and Telegram model-switch validation

PR #97407, opened 2026-08-28
fix(gateway): code-skew guard fingerprints the checked-out tree, not the commit

valicen-davidsaunders:
## What does this PR do?

`gateway/code_skew.py` decides "this process is running stale code" by comparing the checkout's **commit SHA** at boot with the one on disk now. Two commits with an identical tree (a branch switch, an `--allow-empty` commit, a history rewrite) therefore read as skew even though every module on disk is byte-for-byte what the process loaded — and a restart would change nothing. In our deployment a branch flip with an identical tree made the dashboard's `/api/model/options` guard return 503 ("Restart required") for hours.

This PR fingerprints the checked-out **tree** (`git rev-parse HEAD^{tree}`) instead, falling back to the existing ref/SHA reader when git is unavailable. It runs only at boot and on demand (model switch / picker load), never on a hot path.

## Related Issue

Fixes # (no existing issue found — searched PRs/issues for "code skew", "stale code restart required")

## Type of Change

## Changes Made

- `gateway/code_skew.py`: new `_tree_fingerprint()` (subprocess `git rev-parse HEAD^{tree}`, 5s timeout, hex-validated); `_fingerprint()` prefers it, falls back to `_read_git_revision_fingerprint`.
- `tests/test_code_skew.py`: `TestTreeFingerprint` — identical tree across commits is NOT skew; a real content change IS; non-git root returns `None`. Existing tests unchanged (they monkeypatch `_fingerprint`).

## How to Test

1. `pytest tests/test_code_skew.py -q` (13 pass).
2. In a gateway/dashboard checkout: `git commit --allow-empty -m x` → `/model` switch and `/api/model/options` no longer demand a restart.
3. Change a file and commit → the guard still fires.

### Code
- [x] Conventional Commits
- [x] Searched existing PRs
- [x] Only related changes
- [x] `pytest tests/ -q` — relevant suites pass (full-suite run in our env: 1856 passed, 1 pre-existing env-specific failure unrelated to this change)
- [x] Tests added
- [x] Tested on: Debian 13 (Linux 6.12), Python 3.11

- [x] Docstrings updated — N/A otherwise
- [x] Cross-platform: `git` invocation is the same on all platforms; failure falls back to the existing reader

### DIFF
--- a/contributors/emails/david@devon.localdomain
+++ b/contributors/emails/david@devon.localdomain
@@ -1,0 +1,2 @@
+valicen-davidsaunders
+# PR #97407 (Valicen / David Saunders; commit-author email from ops host)
--- a/gateway/code_skew.py
+++ b/gateway/code_skew.py
@@ -23,13 +23,44 @@ _PROJECT_ROOT = Path(__file__).resolve().parent.parent
 _boot_fingerprint: str | None = None
 
 
+def _tree_fingerprint() -> str | None:
+    """Fingerprint of the checked-out TREE (``HEAD^{tree}``), not the commit.
+
+    Two commits with identical content (a branch flip, a rebase that changes
+    nothing, a merge that only rewrites history) must not read as "stale
+    code": the modules on disk are byte-for-byte what the process loaded, so a
+    restart would change nothing. Comparing commit SHAs produced exactly that
+    false positive on 2026-08-27 — a branch switch with an identical tree
+    503'd the dashboard model picker for hours. Spawning ``git`` here is fine:
+    this runs once at boot and then only on demand (model switch / picker
+    load), never on a hot path. Returns ``None`` when git is unavailable so
+    the caller falls back to the ref/SHA reader.
+    """
+    try:
+        import subprocess
+
+        out = subprocess.run(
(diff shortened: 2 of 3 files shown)


valicen-davidsaunders:
Related, for triage context: **#66999** (graceful reload on checkout drift) and issues **#69754** / **#86207** / **#95294** are about what to *do* when skew is real. This PR is narrower and complementary — it stops the guard from firing when skew is *not* real (identical tree under a different commit: branch switch, empty commit, history rewrite). Both can land independently; with #66999 the reload would simply trigger less often.

valicen-davidsaunders:
Pushed a follow-up commit: the fingerprint now hashes the blob ids of non-doc paths from `git ls-tree -r HEAD` instead of the whole tree id, so a docs/README/website-only commit no longer demands a restart either (nothing the process runs changed). Test added for the docs-only case; the identical-tree and real-change cases are unchanged.

valicen-davidsaunders:
Nudge from the Valicen side: this is still relevant on current `main` (we carry it in production on our fork, re-expressed on the post-decomposition tree as of v2026.9.14 — `gateway/code_skew.py` `_tree_fingerprint`). Happy to rebase this PR onto `main` if there is maintainer interest; just say so.

PR #93863, opened 2026-08-24
fix(kanban): run wake turns under the target profile's runtime scope

liuhao1024:
## What does this PR do?

Under gateway multiplexing, a turn started by the **kanban notifier's wake** path ran without the target profile's secret scope (#93851). `deliver_wake` calls `adapter.handle_message` directly, bypassing the inbound dispatch layer that installs the per-turn profile scope — so the woken agent's credential reads (`get_secret('ANTHROPIC_TOKEN')` warning in the reporter's logs) and in-turn adapter/client resolution fell back to the default profile. On the reporter's 9-profile Slack deployment, the default bot is not a member of the target DM, so every in-turn progress send failed `channel_not_found` and the completion report was silently lost (only the final response, re-scoped elsewhere, still delivered). This wraps the push-wake `deliver_wake` call in the same `_profile_runtime_scope` the slash-command path uses for the identical dispatch bypass, resolving the profile home via `_resolve_profile_home_for_source` on the wake's `SessionSource`.

## Related Issue

Fixes #93851

## Type of Change

- [x] 🐛 Bug fix

## Changes Made

- `gateway/kanban_watchers.py` — `_push_wake`: resolve the wake source's profile home and run `deliver_wake` inside `_profile_runtime_scope(home)` (the same wrapper `gateway/slash_commands.py` uses for its dispatch bypass), with a comment documenting the #93851 failure chain. A resolution failure falls back to `nullcontext()` — the unscoped wake, exactly the old behavior — rather than skipping the turn. The notifier's own text ping (already routed via `_authorization_adapter`) and the non-push self-post branch are untouched; single-profile gateways are unaffected (`_resolve_profile_home_for_source` resolves to the default home).
- `tests/hermes_cli/test_kanban_notify.py` — new `test_notifier_wake_runs_under_profile_secret_scope`: a `notify+wake` sub stamped with `notifier_profile="jini"` (routed through `_profile_adapters`, mirroring multiplex wiring) wakes the agent with `current_secret_scope()` active at the moment `deliver_wake` runs; on unpatched `main` the scope is `None` and the test fails.

## How to Test

- New: `.venv/bin/python -m pytest tests/hermes_cli/test_kanban_notify.py::test_notifier_wake_runs_under_profile_secret_scope -q` — should pass. On unpatched `main` it fails (the wake runs with no scope), pinning the regression.
[...]

PR #89276, opened 2026-08-18
fix(gateway): make kanban notifier text pings delivery-reliable (retry + ledger + wake reply anchor)

garadice:
## Symptom

Two delivery-reliability gaps in the kanban notifier (`gateway/kanban_watchers.py`):

1. The task-event **text ping** used a single bare `adapter.send()`. One transient error — e.g. a Telegram FloodWait — permanently failed the delivery, and a gateway crash between a successful send and the cursor advance silently lost the event (the subscription cursor had nothing to resume from). The main agent response path already solves exactly this with `_send_with_retry()` (exponential backoff, FloodWait honor, plain-text fallback) plus the delivery ledger (`gateway/delivery_ledger.py`); the notifier bypassed both.
2. The **wake** `SessionSource` for push-capable adapters was built without `message_id`, even though the subscription's `delivery_metadata` carries `telegram_reply_to_message_id`. The Telegram adapter's routing chain reads `source.message_id` as the reply anchor; when it's `None`, the wake response falls back to the generic `direct_messages_topic_id` and renders outside the visible DM topic lane the user actually reads.

## Fix

- Wrap the text ping in the delivery ledger (`record_obligation` → `mark_attempting` → send via `adapter._send_with_retry(...)` → `mark_delivered` / `mark_failed` in the existing except path). The existing `SendResult(success=False) ⇒ counted as failed` contract is preserved.
- Build the wake `SessionSource` with `message_id=str(telegram_reply_to_message_id)` when the subscription metadata carries one.

## Validation

- New regression tests: the ping observably rides `_send_with_retry` (never a bare `send`) and leaves a `delivered` obligation row in the ledger DB; the wake source for a subscription with `telegram_reply_to_message_id: "462"` carries `message_id == "462"`. Both fail on the previous code.
- Test doubles in the kanban notifier suites updated with minimal `_send_with_retry` stand-ins matching the base-class contract. Full notifier/wake suites: 38 passed across the six files; the broader `-k kanban` gateway selection passes 31/1 skipped (2 pre-existing relay collection errors are unrelated — missing optional dep in this checkout, fail identically on the base tag).

PR #83952, opened 2026-08-11
fix(kanban): make notification delivery durable

whzy1314:
## Summary

- persist per-event/per-part Kanban notification receipts in a durable ledger
- suppress duplicate text and artifact sends during partial-batch retries
- prune receipts older than 90 days once their task is done, archived, or deleted; active-task receipts are retained regardless of age
- support legacy image adapters through a one-image `send_multiple_images` compatibility path while keeping per-file receipt keys
- treat unsuccessful media adapter results as delivery failures so cursors rewind safely
- restrict stale Telegram topic fallback to positively identified private DMs, preventing forum/group media from escaping into the root chat
- retain queryable platform message IDs and delivery timestamps for auditability

## Why

Kanban notification cursors record that an event was claimed, not that every text or media part was delivered. A partial batch failure could therefore resend confirmed parts, while a failed media `SendResult` could be treated as success. Telegram stale-topic media fallback could additionally remove `message_thread_id` outside private DMs.

The durable ledger closes those gaps without growing forever: an hourly watcher sweep uses `kanban.notify_delivery_retention_days` (default 90) to remove old terminal/orphan receipts, including on boards with no remaining subscriptions.

## Verification

- `python3 -m pytest tests/gateway/test_kanban_notifier.py tests/gateway/test_kanban_watchers_mixin.py tests/gateway/test_telegram_thread_fallback.py tests/hermes_cli/test_kanban_notify.py -q` — 78 passed
- regression sabotage confirmed the new retention and legacy-adapter tests fail when their fixes are removed
- `py_compile` passed on changed production and test files
- `git diff --check upstream/main...HEAD` passed
- Codex review of the full diff against current `upstream/main` found no actionable correctness issues

## Notes

This is squashed onto current upstream `main`; no live deployment or real Telegram send is part of this PR.

PR #81143, opened 2026-08-07
fix(gateway): route kanban notices by transport owner

PaulBlackSwan:
## Summary
- preserve the profile that owns the inbound transport separately from the routed runtime profile
- stamp Kanban auto-subscriptions with that transport owner, including shared Relay delivery
- cover routed done/blocked topic notifications and reported-send failure retry semantics

## Verification
- `scripts/run_tests.sh tests/gateway/test_session_context_inheritance.py tests/tools/test_kanban_tools.py tests/gateway/test_multiplex_profile_authz.py tests/gateway/test_kanban_notifier.py -q` (48 passed)
- independent staged-diff review passed with no security or logic findings
- live gateway restarted on this commit; controlled Telegram topic-2 canary completed and its subscription was acknowledged
- replayed the missed blocked event for `t_ba8b0002`; notifier advanced the cursor and woke the subscribed topic

Fixes local Kanban task `t_1c390707`.

PR #97408, opened 2026-08-28
fix(kanban): deliver notify subscriptions for profiles that cannot host the platform

valicen-davidsaunders:
## What does this PR do?

Kanban notify subscriptions are stamped with the profile that created them, and delivery is strictly owner-scoped: only a gateway hosting that profile's adapters may send, with no default-profile fallback (correct — never send from the wrong bot). But in a one-gateway-per-profile deployment where worker profiles run **no** messaging platforms at all, a subscription to e.g. `telegram:<operator DM>` created by a worker profile is unclaimable forever: the owner's gateway skips every tick ("no connected adapters", DEBUG), the platform-hosting gateway skips the row as foreign-owned, and the cursor never advances. No error anywhere; the operator is promised a notification that never arrives.

This PR lets a gateway that hosts the platform deliver on the owner's behalf **only when the owner profile cannot host it anywhere** — decided from configuration (`_profile_hosts_platform()` reads the owner profile's `.env` for the platform's token via `PLATFORM_TOKEN_ENV_NAMES`). Owners that run their own bot are untouched (fail-closed as before); `default`-owned and legacy unstamped rows are never treated as orphans; the zero-subscription fast path only skips a board with no rows at all. Fallback deliveries are **passive-only** (a non-owner gateway must not wake "the agent" in the owner's name), logged at INFO.

## Related Issue

Fixes # (no existing issue found — searched "kanban notify fallback", "notifier profile no adapter")

## Type of Change

## Changes Made

- `gateway/kanban_watchers.py`: `_profile_hosts_platform()`; orphan-candidate collection in `_collect()` (rows owned by non-hosted profiles on platforms this gateway hosts); `fallback` deliveries use this gateway's own adapter and are downgraded to `notify`; zero-sub gate widened.
- `tests/gateway/test_kanban_notifier.py`: fallback delivers (and does not wake); owner-with-own-bot still not served; `.env` parsing.

## How to Test

1. `pytest tests/gateway/test_kanban_notifier.py tests/gateway/test_kanban_notifier_zero_sub_gate.py -q`
2. Live: from a profile with no messaging platforms, `kanban_notify` a Telegram chat the root gateway serves; complete the task → the root gateway logs `delivering ['completed'] for <task> on behalf of profile <p>` and the message arrives. (In our fleet the first tick delivered four notifications that had been stranded for days.)

### Code
- [x] Relevant suites pass (`tests/gateway/test_kanban_notifier*.py`, 34 tests); full suite in our env: 1856 passed
- [x] Tests added
- [x] Tested on: Debian 13, Python 3.11, 15-gateway multi-profile deployment

- [x] Docstrings explain the rule and its limits — N/A otherwise
- [x] Cross-platform: pure Python/`.env` parsing

### DIFF
--- a/contributors/emails/david@devon.localdomain
+++ b/contributors/emails/david@devon.localdomain
@@ -1,0 +1,2 @@
+valicen-davidsaunders
+# PR #97408 (Valicen / David Saunders; commit-author email from ops host)
--- a/gateway/kanban_watchers.py
+++ b/gateway/kanban_watchers.py
@@ -668,6 +668,17 @@ class GatewayKanbanWatchersMixin:
                         sub["chat_id"], sub.get("thread_id") or "",
                     )
                     mode = sub.get("delivery_mode") or "notify"
+                    if d.get("fallback") and mode != "notify":
+                        # A fallback delivery runs on a gateway that is NOT the
+                        # subscription owner's: waking "the agent" here would
+                        # wake this gateway's agent in the owner's name (the
+                        # root agent answered for chief-of-staff tasks on the
+                        # first live run, 2026-08-28). Deliver passively only.
+                        logger.info(
+                            "kanban notifier: fallback delivery for %s downgraded %s -> notify (owner profile %s not hosted here)",
+                            sub["task_id"], mode, sub.get("notifier_profile"),
+                        )
+                        mode = "notify"
                     wake_agent = mode in ("notify+wake", "wake")
                     send_passive = mode != "wake"
                     # Worker handoff carried into the synthetic wake turn below
--- a/gateway/kanban_watchers.py
+++ b/gateway/kanban_watchers.py
@@ -170,6 +170,44 @@ def _release_singleton_lock(handle) -> None:
         pass
 
 
+def _profile_hosts_platform(profile: str, platform_str: str) -> bool:
+    """Does *profile* run its own adapter for *platform_str*?
+
+    Decided from configuration, not from this process's registries: in a
+    one-gateway-per-profile deployment the root gateway cannot see what other
+    gateway processes host, but it CAN read the owner profile's ``.env``. A
+    profile that has no bot token for the platform cannot deliver on it
+    anywhere, so refusing to deliver on its behalf just drops the message —
+    the "kanban_notify to telegram from a worker profile silently never
+    arrives" failure (Valicen, 2026-08-27). Unknown platforms (no token env
+    var in the canonical map) answer True: fail closed, no fallback.
+    """
+    try:
+        from gateway.config import PLATFORM_TOKEN_ENV_NAMES, Platform as _P
+        from hermes_cli.profiles import get_profile_dir
+
+        try:
+            plat = _P(platform_str)
+        except ValueError:
+            return True
+        env_name = PLATFORM_TOKEN_ENV_NAMES.get(plat)
+        if not env_name:
+            return True
+        env_file = get_profile_dir(profile or "default") / ".env"
+        if not env_file.is_file():
+            return False
+        for line in env_file.read_text(encoding="utf-8", errors="replace").splitlines():
+            line = line.strip()
+            if line.startswith("export "):
+                line = line[7:].strip()
+            key, sep, value = line.partition("=")
+            if sep and key.strip() == env_name:
+                return bool(value.strip().strip("'\""))
+        return False
+    except Exception:
+        return True
+
+
 def _wake_scope_id(adapter: Any, sub: dict) -> Optional[str]:
     """Return the tenant scope (Slack workspace) a subscription's wake keys to.
 
@@ -401,10 +439,14 @@ class GatewayKanbanWatchersMixin:
                                 board=slug,
                                 notifier_profiles=notifier_profiles,
                                 include_unowned=include_unowned,
-                            ) == 0:
+                            ) == 0 and _kb.count_notify_subs(board=slug) == 0:
+                                # Both owned AND total are zero. A non-zero
+                                # total with zero owned still needs the open:
+                                # the delivery fallback below may serve rows
+                                # whose owner profile cannot host the platform.
                                 logger.debug(
-                                    "kanban notifier: board %s has no subscriptions owned by %s; skipping open",
-                                    slug, sorted(notifier_profiles),
+                                    "kanban notifier: board %s has no subscriptions at all; skipping open",
+                                    slug,
                                 )
                                 continue
                         except Exception as exc:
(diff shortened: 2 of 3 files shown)


valicen-davidsaunders:
Related upstream work, for triage context (not duplicates — they fix the *stamping* side; this PR fixes the *delivery* side):

- **#81143** stamps auto-subscriptions with the transport-owning profile (`HERMES_SESSION_TRANSPORT_PROFILE`). That prevents mis-stamping for sessions that arrived over a transport. It does not cover the case here: a subscription created from a session with no inbound transport (dashboard/browser chat, CLI) targeting a platform only another gateway hosts — the row is stamped with a profile that *cannot* host that platform anywhere.
- **#76514** / issue **#76483** fix the CLI stamping the invoking shell's profile — same symptom family ("named-profile gateway skips it forever").

This PR is complementary: whatever profile a row is stamped with, if that profile has no token for the platform (checked from its `.env`), the gateway that does host the platform delivers — passive-only, and never for owners that run their own bot. With #81143 merged, fewer rows would need the fallback; existing mis-stamped rows and transport-less sessions still would.

andrexibiza (commented):
Reviewed exact head `d383bd33467368805b8a3e86584d7d8f4633a706` against live `main@e60983a69730c058ce772829df3273aee6de3889` (merge base `00bbfc690060d1323ddb2f065297c7425cb71c26`; this branch is 2 commits ahead / 48 behind). I read the current notifier/config/authz paths, both commits, the focused tests, hosted CI, and the overlapping Kanban-notification work. There was no existing submitted review on this head.

The second commit's **passive-only** containment is the right instinct: a gateway borrowing another transport must not also wake an agent under the wrong profile. The live stranded-notification evidence is useful too. I do think two P1 authority/routing gaps need to close before this can safely become the delivery fallback, though.

### P1 — `.env` absence is not proof that the owner cannot host the platform

`_profile_hosts_platform()` turns “token not present in `<profile>/.env`” (and even “profile `.env` does not exist”) into `False`, and that `False` is the authorization for another gateway to deliver the subscription.

That is weaker than Hermes' actual credential source of truth. On current main, `gateway/config.py::_getenv()` first reads the active profile's `secret_scope` and otherwise reads `os.environ`; Telegram/Discord/Slack/Mattermost/Matrix token loading goes through that path. A profile can therefore have a real, live platform credential with no token in its `.env` at all. From another gateway process there is no way to prove “cannot host anywhere” by inspecting only that file.

The consequence is not just a stale heuristic: it widens an explicitly protected boundary. `GatewayAuthorizationMixin._authorization_adapter()` deliberately refuses to fall back from a stamped secondary profile to the default profile when that profile has no live registry entry, specifically because doing so can send through the wrong bot. This PR re-opens that fallback based on a non-authoritative absence signal.

**Required:** do not let `.env` absence mint delivery authority. Resolve the owner's capability from the same authoritative profile/config/secret mechanism used at adapter admission, or (preferably) stop inferring it and persist the exact transport/delivery owner on the subscription. Unknown/unreadable/unobservable owner capability must remain fail-closed. Please add the adversarial witness where the owner has no `.env` token but does have a token via scoped/process credential resolution; another gateway must not take the subscription.

### P1 — the fallback selects a platform, not an exact transport identity

The fallback candidate is qualified only by the platform string, then delivery uses this process's `self.adapters[plat]`. In a multi-profile/fleet deployment there may be several Telegram/Slack/etc. adapters with different bot/workspace/account identities. If more than one gateway decides the owner “cannot host,” the first process to win `claim_unseen_events_for_sub()` effectively chooses the sending identity by scheduling race rather than by durable subscription identity.

There is a concrete multiplex variant in this diff too: `_collect()`'s `active_platforms` intentionally includes platforms hosted only in `_profile_adapters`, but fallback delivery looks only in `self.adapters`. A process can therefore classify itself as hosting the platform, claim the row, and then have no fallback adapter at the actual send boundary.

This is exactly where the adjacent transport-owner work matters:

- #81143 (`PaulBlackSwan`) preserves the **transport-owning profile separately from the routed runtime profile** and stamps Kanban auto-subscriptions with it.
- #76514 (`chelsealong`) fixes the CLI side by refusing to pretend the invoking shell profile is the serving notifier profile when it cannot know that.

Those are complementary, not duplicates of this repair. I would reconcile them first, then define the residual case as “subscription has no resolvable delivery owner” rather than “owner profile lacks a token in one file.” If a residual fallback still exists, it needs one exact adapter/account/tenant identity at claim/send time, not `any gateway with this Platform`. Please add a two-gateway/same-platform/different-account regression proving only the intended transport can claim/send, plus the secondary-`_profile_adapters`-only case.

### Landing graph / settlement

There is active same-file work that needs an explicit restack rather than independent ownership of the notifier loop: #83952 (`whzy1314`) adds durable per-part notification receipts, #89276 (`garadice`) adds retry + delivery-ledger settlement, #93863 (`liuhao1024`) is the wake/profile-scope side and contains real wrong-profile Slack evidence, and #75281 (`xaviersudre`) currently changes both `gateway/kanban_watchers.py` and `tests/gateway/test_kanban_notifier.py` for durable workflow subscriptions. These are adjacent/complementary classes; none should lose credit or be flattened into “duplicate notifier work.” The safe order is: establish canonical subscription/transport identity, then attach durable delivery settlement/retry to that exact route, then keep this fallback only for the residual identity state the first step cannot resolve.

### Acceptance / attribution

The exact head does not have hosted acceptance yet: CI `33206268120`, Docker `33206267138`, and Nix `33206267251` are `action_required`; the CI run has **zero jobs**. The first commit `3e2918eda0aa4cffdea8fab66dc70bde68744f20` has no workflow runs/check-runs at all. The reported 34 focused / 1856 local tests are useful development evidence, but they do not make either commit hosted-green.

Both commits are authored `David Saunders <david@devon.localdomain>` and preserve their cherry-pick source SHAs, but current main has no contributor mapping for that email to `valicen-davidsaunders`. Please preserve the existing authorship/cherry-pick lineage and add the release-resolvable mapping rather than rewriting credit away.

The useful core here is real: stranded subscriptions should not stay silent forever, `default`/legacy-unowned rows are intentionally excluded, and passive-only fallback prevents a cross-profile wake. Once the decision to cross the owner boundary is based on qualified transport identity rather than `.env` absence and claim timing, this becomes a much stronger repair.

valicen-davidsaunders:
Thanks for the thorough review — agreed on both P1s.

- **Send boundary bug** (`_collect()` counting `_profile_adapters`-only platforms as hosted while the fallback send site uses `self.adapters`): fixed on our side by qualifying fallback candidates by `self.adapters` only, with a regression test (`test_no_fallback_claim_when_platform_hosted_only_by_secondary_profile`). I'll fold that into this branch when it's restacked.
- **Owner capability from `.env` absence**: you're right that it's not the authoritative credential path. Rather than bolt a second resolution mechanism onto this PR, I'll restack once #81143 (transport-owning profile on the subscription) and #76514 land, and redefine the residual case as "subscription has no resolvable delivery owner" with an exact adapter identity at claim/send time, plus the two-gateway/same-platform/different-account and `_profile_adapters`-only witnesses you asked for.

Converting to draft until then. Attribution mapping for `david@devon.localdomain` was added in 95fde594 (no history rewrite).

smshd:
Supporting data point for this PR. We tried to do owner fallback as a plugin on `kanban_task_completed` / `kanban_task_blocked` and hit a wall: those hooks fire in the worker process and carry no delivery outcome, while the notifier and its wake injection run in the gateway. So a plugin can only add a parallel notification path, not a true fallback. That is why a proper fallback needs a core change like this one. Our plugin, with that limitation stated up front: https://github.com/Smshd-Avo/hermes-kanban-owner-fallback

valicen-davidsaunders:
@smshd thanks — that is exactly the limitation we hit too; a plugin can add a parallel path but not a fallback. Status from our side: the send-boundary P1 (fallback candidates qualified by `self.adapters` only, with a regression test) is fixed on our fork and in production since 2026-08-28, and the branch is re-homed onto `gateway/kanban_watchers_notifier.py` alongside the multiplex-aware `_adapter_for_subscription` that landed upstream in September. Still parked as draft pending #81143 / #76514 to restack on — maintainers, if you would rather take it as-is with the `.env` heuristic replaced by the resolver you prefer, say which and I will un-draft and rebase.

PR #50288, opened 2026-06-21
fix(kanban): treat research worker's comment-then-exit as completed, not protocol violation

JonT6:
## Problem

When a research/audit worker (researcher profile) ships a result as a
`kanban_comment` (a substantive handoff with file paths) then exits
cleanly with rc=0, the dispatcher flags the run as a "protocol
violation" and **respawns a duplicate worker**. The orchestrator gets no
completion notification.

Reproduced 2026-06-21 on a cheap-NO audit (task `t_c928fc79`):
- Run 158 posted a 2325-byte handoff comment with a report path, exited rc=0
- Dispatcher marked the run as `crashed` with reason "worker exited cleanly (rc=0) without calling kanban_complete or kanban_block — protocol violation"
- Run 159 (duplicate) was spawned before the orchestrator could intervene

This is incorrect classification. A clean exit accompanied by a
substantive handoff comment **is** a successful completion for a research
task. The protocol-violation detector should not punish the
comment-then-exit pattern.

## Fix

Patch `detect_crashed_workers` in `hermes_cli/kanban_db.py` to inspect
the worker's last comment when classifying a clean exit. If a comment
exists that:

1. Was created at or after the task's `started_at` (scope: not stale)
2. Has length >= 500 bytes (substantive)
3. References a durable artifact (regex match for `data/...`, `docs/...`,
   `/tmp/...`, `.md`/`.json`/`.txt` filenames, `https?://` URLs, or the
   phrase `pull request`)

...then the worker is auto-promoted to `done` with event
`commented_through_completion`, **not** marked as a protocol violation
and **not** respawned.

Otherwise (no comment, comment < 500 bytes, or no artifact ref), the
existing protocol-violation behavior is preserved — silent failures
should still be flagged.

## TDD

5 new tests in `tests/hermes_cli/test_kanban_core_functionality.py`:

- `test_clean_exit_with_substantive_comment_marks_done` (RED → GREEN)
- `test_clean_exit_with_short_comment_still_protocol_violation` (pin: unchanged)
- `test_clean_exit_with_no_comment_still_protocol_violation` (pin: unchanged)
- `test_clean_exit_with_artifact_in_500byte_comment_marks_done` (RED → GREEN)
- `test_clean_exit_comment_from_prior_run_does_not_save_it` (pin: stale-comment scoping)

Red step: 2 of 5 failed on `origin/main` before the patch. Green step:
all 5 pass after. 8 related protocol-violation / circuit-breaker tests
continue to pass. No regressions.

[...]

PR #97409, opened 2026-08-28
fix(kanban): route comment-then-clean-exit workers to review instead of retry

valicen-davidsaunders:
## What does this PR do?

A worker that finishes its task, posts the result as a comment on the card, and then exits `rc=0` **without** calling `kanban_complete` is today treated as a protocol violation: the dispatcher retries, the retry redoes finished work, exhausts the iteration budget, and the task ends up auto-blocked with `Iteration budget exhausted` — a **done** task reported as failed (we hit this repeatedly; the card's comments were the ground truth, the failure string was wrong).

This PR makes `detect_crashed_workers` check for a comment by the assignee posted during the run. If one exists, the exit is an `unconfirmed_completion`: the run closes with that outcome, no failure is counted, `last_failure_error` is cleared, and the task is handed to the review phase via `request_review(force=True, summary=<the worker's comment>)`, which emits `review_requested` so the subscriber is woken to confirm (`kanban_complete`) or send it back. Clean exits **without** such a comment are handled exactly as before (protocol violation, bounded retry).

## Related Issue

Fixes # (no existing issue found — searched "protocol violation completion comment review")

## Type of Change

## Changes Made

- `hermes_cli/kanban_db.py` (`detect_crashed_workers`): comment lookup for clean exits; `unconfirmed_completion` event + run outcome; post-txn `request_review` handoff with warnings on refusal.
- `tests/hermes_cli/test_kanban_unconfirmed_completion.py`: with comment → `review`, no failure counted, events/outcomes as expected; without comment → protocol violation unchanged.

## How to Test

1. `pytest tests/hermes_cli/test_kanban_unconfirmed_completion.py tests/hermes_cli/test_kanban_db.py -q`
2. Live: let a worker post a comment and exit without a terminal call → the task lands in `review` with the comment as summary and the origin is notified.

### Code
- [x] Relevant suites pass (kanban db/review/lifecycle/notifier: 99 tests); full suite in our env: 1856 passed
- [x] Tests added
- [x] Tested on: Debian 13, Python 3.11

- [x] Docstring/comments updated — N/A otherwise
- [x] Cross-platform: N/A (SQLite + Python only)

### DIFF
--- a/contributors/emails/david@devon.localdomain
+++ b/contributors/emails/david@devon.localdomain
@@ -1,0 +1,2 @@
+valicen-davidsaunders
+# PR #97409 (Valicen / David Saunders; commit-author email from ops host)
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -8901,6 +8901,7 @@ def detect_crashed_workers(conn: sqlite3.Connection) -> list[str]:
     # Worker-exit observer payloads (RFC #58548), collected inside the main
     # txn and fired only after every reclaim/accounting txn has committed.
     exited_hook_payloads: list[dict] = []
+    review_candidates: list[tuple] = []  # (task_id, comment_excerpt, run_id)
     with write_txn(conn):
         rows = conn.execute(
             "SELECT id, worker_pid, claim_lock, started_at, assignee "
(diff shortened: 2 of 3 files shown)


valicen-davidsaunders:
Prior art I missed on the first search: **#50288** (June) targets the same trigger — a worker posts a substantive comment during the run and exits rc=0. The two differ in what happens next, and I think that difference is the point worth deciding on:

- #50288 classifies the run as **completed → `done`**, gated by heuristics (≥500 bytes and an artifact-looking pattern in the comment).
- This PR classifies it as an **unconfirmed completion → `review`** (via `request_review`, which emits `review_requested` and wakes the subscriber), with no size/regex heuristics — only "a comment by the assignee posted during this run". A human or reviewer confirms with `kanban_complete` or sends it back.

Routing to review keeps the fail-closed property (nothing is auto-marked done on a guess) while still ending the retry/auto-block loop, and it reuses the existing review phase rather than adding a completion heuristic. Happy to fold this into #50288 (or the reverse) if maintainers prefer one PR — whichever shape you want.

valicen-davidsaunders:
Nudge: still relevant on current `main` — we run it in production on our fork, re-homed onto the `_classify_dead_worker` wrapper / `_classify_dead_worker_exit` split that landed in September (v2026.9.14). The open design question from my last comment (route to review vs #50288's auto-complete) is the one worth a maintainer call; I will rebase onto `main` on request.

PR #97395, opened 2026-08-28
fix(bot-mode): preserve canonical capabilities after compression

helix4u:
## What does this PR do?

Current Hermes compresses sessions in place by default, preserving the same session ID and title. Hermes also supports existing compression lineages created by the legacy rotation path or by installs configured with `compression.in_place: false`. In those lineages, the canonical `Bot Chat` is the root session while the active continuation can have a different display title.

Several Bot Mode capability gates checked only the active session's title, so a valid compression descendant of `Bot Chat` could lose the Bot Mode protocol and `message_agent` capability. The diagnostic established that the affected session belonged to such a lineage, but did not include the original compression event, so this PR does not assume how or when that lineage was created.

This centralizes the identity check and resolves compression lineage back to its canonical root. It deliberately uses the compression-only lineage contract, which excludes explicit branches, delegated agents, and tool sessions. Normal in-place compression behavior is unchanged.

## Related Issue

No public issue. Reproduced from a support diagnostic and current source.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Add one lineage-aware Bot Chat identity helper in `tools/bot_mode_probe.py`.
- Use that helper for system-prompt protocol injection, capability-epoch refresh, TUI capability sync, and both `message_agent` gates.
- Add regressions proving a renamed compression descendant retains Bot Chat capabilities while a delegated child does not inherit them.

## How to Test

1. Create a canonical session titled `Bot Chat`, then model an existing compression-rotation lineage with a continuation whose display title differs.
2. Resolve Bot Mode identity and inject `message_agent` against the continuation.
3. Confirm the continuation uses the root identity, while an explicit delegate child remains outside Bot Mode.

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
[...]

PR #97411, opened 2026-08-28
fix(bot-mode): compressed Bot Chats keep their protocol and message_agent (salvage #97395)

teknium1:
## Summary

A compressed Bot Chat no longer loses its Bot Mode protocol and `message_agent` — the capability gate now recognizes the canonical `Bot Chat` title anywhere in the session's compression lineage instead of only on the live row's title. Salvages #97395 by @helix4u with an inverted-premise correction on top.

Root cause of the field report (support diagnostic `8f46d199`): compression rotates onto a new session row, and when the best-effort title carry fails (it is wrapped in a swallow-all `except`) or the rows predate the title-carry contract, the canonical title strands on a hidden ancestor while the live tip shows a generated title. Every title-gated Bot Mode site then saw a non-Bot-Chat session and silently dropped the protocol and `message_agent`.

**Correction over the original PR:** #97395 gated on the lineage ROOT's title. But the healthy rotation contract transfers `Bot Chat` off the hidden ancestor ONTO the tip (`_set_session_title`'s compression-ancestor transfer resolves the UNIQUE(title) conflict), so after a normal compaction the root's title is NULL — a root-only check would have dropped capabilities on every healthy compressed Bot Chat. The gate now scans every lineage member, covering both the healthy (tip-titled) and degraded (ancestor-stranded) shapes; delegates/branches/tool sessions remain excluded by the `get_compression_lineage` contract.

## Changes

- `tools/bot_mode_probe.py`: new `is_bot_chat_session()` — whole-lineage title scan; a live row that already carries a DIFFERENT title beats a stale `Bot Chat` hint; untitled fresh rows still fall back to the hint.
- `tools/bot_mode_dm.py`, `agent/system_prompt.py`, `agent/conversation_loop.py`, `tui_gateway/server.py`: all five title-gate sites (prompt injection, stale-prompt refresh, TUI capability sync, both `message_agent` gates) route through the one helper; duplicate `_session_title()` deleted.
- Tests: healthy title-carried-to-tip, real `publish_compression_child` + title-carry regression, stranded-ancestor recovery, delegate exclusion, stale-hint rejection, untitled-hint fallback.

## Validation

| Check | Result |
|---|---|
| `tests/tools/test_bot_mode_probe.py` + `test_bot_mode_dm.py` + bot/compression subset | 130 passed, 2 skipped |
| Sabotage: PR's root-only logic | fails `healthy_title_carried_to_tip` (proves the inversion) |
| Sabotage: main's tip-only logic | fails `stranded_on_ancestor` + `stale_hint` (proves the fix) |
| ruff + Windows footguns | clean |

Cache-safe: lineage titles are stable for a session's life; the gate is deterministic per turn, tool list stays byte-identical.

Credit: @helix4u for the diagnosis, the helper centralization, and the lineage approach — cherry-picked with authorship preserved; the lineage-scan correction is a follow-up commit on top.

## Infographic

Bot Chat survives compaction

### DIFF
--- a/agent/conversation_loop.py
+++ b/agent/conversation_loop.py
@@ -929,6 +929,7 @@ def _restore_or_build_system_prompt(agent, system_message, conversation_history)
         try:
             from tools.bot_mode_probe import (
                 BOT_CHAT_TITLE,
+                is_bot_chat_session,
                 stored_bot_chat_prompt_needs_upgrade,
                 stored_prompt_capability_stale,
             )
@@ -948,14 +949,10 @@ def _restore_or_build_system_prompt(agent, system_message, conversation_history)
                 # the messaging protocol. Title-gated so ordinary unstamped
                 # sessions (i.e. all of them) never take this path; the
                 # rebuilt prompt carries the stamp, so it cannot re-fire.
-                _t = str(getattr(agent, "_session_title_hint", "") or "").strip()
-                if not _t and agent._session_db and agent.session_id:
-                    try:
-                        _t = str(agent._session_db.get_session_title(agent.session_id) or "").strip()
-                    except Exception:
-                        _t = ""
-                if _t == BOT_CHAT_TITLE:
-                    _bot_stale = stored_bot_chat_prompt_needs_upgrade(stored_prompt, _home_for_epoch)
+                if is_bot_chat_session(agent):
+                    _bot_stale = stored_bot_chat_prompt_needs_upgrade(
+                        stored_prompt, _home_for_epoch
+                    )
         except Exception:
             _bot_stale = False
         if _bot_stale:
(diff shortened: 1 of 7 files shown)


teknium1:
Closing: the diagnostic bundle this was built against shows **no compression event at all** — the reported session (`20260826_225535_ae7751`) keeps one session id from creation through the latest repro, so the "capabilities lost after compaction" premise (both #97395's and this salvage's) is not supported by the field report. Credit to @helix4u for re-tracing the actual path (session-selection via `resolve_resume_session_id()`'s legacy child-walk).

The lineage-scan gate here is shelved unless a real stranded-title report shows up. The separate log-proven bug from the same bundle (deleted `newbottesting` profile resurrected by the desktop backend pool) is being fixed independently.

PR #1, opened 2025-07-26
Terminal tool

hjc-puro:
image

```sh
python run_agent.py \
  --query "search up the latest docs python 3.13 and write me an example using the new jit compiler that's not in the docs. profile its performance and present the results in a beautiful markdown table right in the terminal." \
  --max_turns 20 \
  --model claude-sonnet-4-20250514 \
  --base_url https://api.anthropic.com/v1/ \
  --api_key $ANTHROPIC_API_KEY
```

PR #83324, opened 2026-08-10
fix(compressor): thread custom_providers into context-length resolution

dhruvkej9:
## Problem

`ContextCompressor._resolve_context_length` called `get_model_context_length` **without** `custom_providers`. As a result, a per-model `context_length` override declared in `custom_providers` (e.g. `256000` for `baseten-deepseek-flash`) was skipped, and the hardcoded catalog default (`128000` for the `deepseek` family) won instead.

Symptom: `/context` reports the wrong context window (128K) even though the user configured 256K for their model. The same override is honored on the `/model` switch path (`agent_runtime_helpers`) and in `agent_init`'s config resolution, but the compressor's own deferred resolution path was the one gap.

## Root cause

`agent/context_compressor.py` — `_resolve_context_length()` (deferred, first-access resolution) did not forward `custom_providers` to `get_model_context_length`, so step 0b of that function's resolution order (custom_providers per-model override) never ran.

## Fix

- Add a `custom_providers` parameter to `ContextCompressor.__init__` and store it.
- Thread it into `get_model_context_length` in `_resolve_context_length`.
- Pass `_custom_providers` from `agent_init.py` when constructing the compressor.

## Verification

- With `custom_providers` threaded: compressor resolves **256000** ✓
- Without (old behavior): resolves **128000** (the bug)
- `tests/hermes_cli/test_custom_provider_context_length.py` + `tests/run_agent/test_invalid_context_length_warning.py`: **9/9 pass**

PR #97412, opened 2026-08-28
fix(metadata): normalize version separators in catalog substring match

### LINKED ISSUE #97398: Context length falls back to hardcoded 202,752 for model slugs with hyphens (e.g. z-ai-glm-5-3) despite custom_providers override and catalog entry
## Bug Description

When using a custom OpenAI-compatible provider with the model slug `z-ai-glm-5-3`, the
agent resolves the model's context window to a hardcoded fallback of **202,752 tokens**
instead of the correct ~1M, even though:

- the internal catalog (`DEFAULT_CONTEXT_LENGTHS` in `agent/model_metadata.py`) contains
  an entry `"glm-5.3": 1_048_576` for exactly this model, and
- the user configured `context_length: 1000000` for this model under
  `custom_providers`.

The desktop app's Context Usage meter then shows a wrong max (e.g. `202.8k`), and
compression thresholds are computed against the wrong window.

Log evidence (agent.log):

```
INFO agent.model_metadata: Could not detect context length for model 'z-ai-glm-5-3'
  at https://api.venice.ai/api/v1 — defaulting to 256,000 tokens (probe-down).
INFO agent.model_metadata: Using hardcoded context length 202,752 for model
  'z-ai-glm-5-3' (custom endpoint, catalog match on 'glm')
```

### 1. Substring catalog match fails on hyphenated slugs

The catalog lookup in `agent/model_metadata.py` (around the `DEFAULT_CONTEXT_LENGTHS`
loop, ~line 3258–3273) lowercases the model name and does a plain substring match:

```python
if default_model in model_lower
```

against keys sorted longest-first. The catalog key is `glm-5.3` (with a **period**),
but the runtime slug is `z-ai-glm-5-3` (hyphenated). `"glm-5.3" in "z-ai-glm-5-3"` is
`False`, so matching falls through to the generic `"glm"` entry → 202,752.

Notably, the same file already handles other hyphen/period variants, so this seems to
be an oversight for the GLM-5.3 entry (and potentially other dotted-version keys vs.
hyphenated provider slugs in general — e.g. any `vendor-model-5-3` style slug will miss
a `model-5.3` key).

**Suggested fix:** normalize separators before matching (e.g. compare with `.`/`-`/`_`
folded to one canonical separator), or add explicit hyphenated aliases.

### 2. `custom_providers` context_length override not threaded to the compressor

`ContextCompressor._resolve_context_length()` (in `agent/context_compressor.py`, ~line
2245) calls `get_model_context_length()` **without** passing the `custom_providers`
list, so `get_custom_provider_context_length()` never sees the user's per-model
`context_length` override.

Verified by direct reproduction:

```python

# with custom_providers threaded through -> 1000000  (correct)
# without                               -> 202752    (fallback)
```

`hermes_cli/config.py::get_custom_provider_context_length` itself works correctly —
the value is simply not passed along on this code path.

**Suggested fix:** thread the custom provider entries (or the resolved context length)
into `ContextCompressor`, mirroring what the startup path in `run_agent.py` already
does via `config_context_length`.

## Steps to Reproduce

1. Configure a custom OpenAI-compatible provider (e.g. Venice, `api.venice.ai/api/v1`)
   with model `z-ai-glm-5-3` and a per-model `context_length: 1000000`.
2. Ensure the provider's `/models` endpoint does not report a context window (Venice
   does not — this triggers the "probe-down" path).
3. Start a session and open the Context Usage popover (or check the status bar).
4. Observe max context `202.8k` instead of `1.0M`.

## Expected Behavior

- Catalog match should recognize `z-ai-glm-5-3` as GLM-5.3 → 1,048,576.
- Independently, an explicit `context_length` under `custom_providers` should win over
  any fallback.

## Actual Behavior

Context window resolves to the generic `"glm"` fallback (202,752) and the
`custom_providers` override is ignored on the compressor code path.

## Workaround

Setting the top-level `model.context_length` in config.yaml takes precedence and
restores the correct window (effective on the next message, no restart needed).

## Environment

- Hermes Agent gateway version 0.20.6 (desktop app, packaged build)
- macOS
- Custom provider: Venice AI (`api.venice.ai/api/v1`), chat_completions mode

shellybotmoyer:
## Summary

Fixes bug #1 from issue #97398: the `DEFAULT_CONTEXT_LENGTHS` catalog substring match fails when a catalog key uses a dot version separator (e.g. `"glm-5.3"`) but the runtime model slug uses hyphens (e.g. `"z-ai-glm-5-3"`).

The plain substring check `"glm-5.3" in "z-ai-glm-5-3"` is `False`, so the model silently falls back to the shorter catch-all entry `"glm"` (202,752 tokens) instead of the correct 1,048,576.

### Fix

Added a `_catalog_key_matches()` helper that normalizes version separators on both sides via the existing `_normalize_model_version()` function (which replaces `.` with `-`) before performing the substring check. Applied at all three `DEFAULT_CONTEXT_LENGTHS` lookup sites:

1. `_stale_pre_catalog_cache_entry()` — stale-cache detection
2. `get_model_context_length()` step 3b — custom endpoint catalog fallback
3. `get_model_context_length()` step 8 — generic hardcoded defaults

This also benefits any other dotted catalog key vs hyphenated slug mismatch (not just GLM).

### Verification

```python
>>> _catalog_key_matches("glm-5.3", "z-ai-glm-5-3")
True        # was False before fix
>>> _catalog_key_matches("glm-5.2", "z-ai-glm-5-2")
True        # existing behavior preserved
>>> _catalog_key_matches("claude-opus-4.8", "claude-opus-4-8")
True        # existing behavior preserved
>>> _catalog_key_matches("claude-opus-4-8", "glm-5-3")
False       # non-matches still correct
```

### Note

This addresses bug #1 (substring catalog match) from #97398. Bug #2 (custom_providers context_length override not threaded to the compressor) is tracked separately in PR #83324.

Closes #97398 (partially — bug #1 only).

### DIFF
--- a/agent/model_metadata.py
+++ b/agent/model_metadata.py
@@ -2227,7 +2227,7 @@ def _stale_pre_catalog_cache_entry(model: str, cached: int) -> bool:
     matches = [
         (key, value)
         for key, value in DEFAULT_CONTEXT_LENGTHS.items()
-        if key in model_lower
+        if _catalog_key_matches(key, model_lower)
     ]
     if not matches:
         return False
@@ -2439,6 +2439,18 @@ def _normalize_model_version(model: str) -> str:
     Normalize both to dashes for comparison.
     """
     return model.replace(".", "-")
+
+
+def _catalog_key_matches(catalog_key: str, model_lower: str) -> bool:
+    """Check if a *catalog_key* is a substring of *model_lower*, normalizing
+    version separators first.
+
+    Catalog keys may use dots (e.g. ``"glm-5.3"``) while provider slugs use
+    hyphens (e.g. ``"z-ai-glm-5-3"``).  Without normalization the substring
+    check ``"glm-5.3" in "z-ai-glm-5-3"`` is ``False`` and the model silently
+    falls back to a shorter catch-all entry (e.g. ``"glm"`` → 202 752).
+    """
+    return _normalize_model_version(catalog_key) in _normalize_model_version(model_lower)
 
 
 def _query_anthropic_context_length(model: str, base_url: str, api_key: str) -> Optional[int]:
@@ -3263,7 +3275,7 @@ def get_model_context_length(
                 key=lambda x: len(x[0]),
                 reverse=True,
             ):
-                if default_model in model_lower:
+                if _catalog_key_matches(default_model, model_lower):
                     logger.info(
                         "Using hardcoded context length %s for model %r "
                         "(custom endpoint, catalog match on %r)",
@@ -3443,7 +3455,7 @@ def get_model_context_length(
     for default_model, length in sorted(
         DEFAULT_CONTEXT_LENGTHS.items(), key=lambda x: len(x[0]), reverse=True
     ):
-        if default_model in model_lower:
+        if _catalog_key_matches(default_model, model_lower):
             return length
 
     # 9. Default fallback — warn (deduped per model+endpoint) so


teknium1:
#106976

PR #16375, opened 2026-04-27
Fix profile-scoped gateway PID detection

kinsolee:
## Summary

- Scope `_get_service_pids()` to the current Hermes profile by default.
- Preserve cross-profile service PID discovery for both update sweeps via `all_profiles=True`.
- Add regression coverage for scoped and all-profile systemd discovery in the current `tests/hermes_cli/test_gateway.py`.

## Why

`hermes cron status` uses `find_gateway_pids()` to decide whether cron jobs will fire automatically. On hosts with multiple Hermes profile services, such as `hermes-gateway.service` and `hermes-gateway-coder.service`, the previous systemd lookup matched `hermes-gateway*` and could report another profile's gateway as the current profile's gateway.

This could produce misleading cron status output, including a false positive when the current profile's gateway is stopped but another profile's gateway is running.

## Verification

- `scripts/run_tests.sh tests/hermes_cli/test_gateway.py -q` — 56 passed
- `.venv/bin/python -m py_compile hermes_cli/gateway.py hermes_cli/main.py`
- `.venv/bin/ruff check hermes_cli/gateway.py hermes_cli/main.py tests/hermes_cli/test_gateway.py`

A full-suite run was also attempted after syncing the branch to current `main`; it was stopped after 3,300+ passing tests because of unrelated environment/main failures (`agent-client-protocol` missing from the initial dev-only environment and an existing LSP live-system-guard failure).

PR #97428, opened 2026-08-28
fix(gateway): scope systemd service PIDs to current profile in gateway status

### LINKED ISSUE #97360: hermes -p <profile> gateway status reports other profiles PIDs and claims "running" when that profile has no gateway
## Summary

`hermes -p <profile> gateway status` reports a profile as running and quotes PIDs that belong to *other* profiles' gateways, even when the named profile has no service installed and has never run one. `hermes gateway list` reports the same profile correctly as not running.

## Reproduction

On a multi-profile install where `default` and `fte-manager` have gateway services running and `it-ops` has none:

```
$ hermes -p it-ops gateway status
✓ Gateway is running (PID: 2081267, 2061253)
  (Running manually, not as a system service)

To install as a service:
  hermes gateway install
  sudo hermes gateway install --system

Other profiles:
  ✓ default          — PID 2081267
  ✓ fte-manager      — PID 2061253
```

PIDs 2081267 and 2061253 are the `default` and `fte-manager` gateways — as the command's own "Other profiles" section states two lines further down. Neither belongs to `it-ops`.

`gateway list`, on the same machine at the same moment, is correct:

```
$ hermes gateway list
Gateways:
  ✓ default (current)        — PID 2081267
  ✗ academy-ops              — not running
  ✗ academy-sales            — not running
  ✓ fte-manager              — PID 2061253
  ✗ it-ops                   — not running
```

`it-ops` genuinely had no gateway — no unit, and no journal history at all:

```
$ ls ~/.config/systemd/user/ | grep hermes
hermes-gateway-fte-manager.service
hermes-gateway.service

$ journalctl --user -u hermes-gateway-it-ops.service
-- No entries --
```

After `hermes -p it-ops gateway install`, a unit is created and `gateway list` flips to `✓ it-ops — PID …`, confirming nothing had been running before.

## Expected

`gateway status` scoped to a profile should report only that profile's gateway, and should say "not running" when the profile has no service — consistent with `gateway list`.

## Actual

It reports "running" and attributes other profiles' PIDs to the named profile. The output also contradicts itself within a single invocation: the headline claims the profile is running while "Other profiles" lists those same PIDs under different profiles.

## Impact

A missing gateway looks healthy. In our case the affected profile was the only one on the machine with a messaging platform configured (a Telegram bot token in its profile `.env`), so its bot was simply never connected — while `gateway status` reported it running every time it was checked. The `(Running manually, not as a system service)` line compounds this by implying a deliberate foreground run rather than an absent service.

## Workaround

Use `hermes gateway list`, which resolves per-profile state correctly.

## Environment

- Hermes Agent v0.20.6
- Linux (Ubuntu), systemd user units, non-root install
- Profiles: default, academy-ops, academy-sales, fte-manager, it-ops

wooyongbin3-cpu:
## Summary

This PR fixes #97360 by properly scoping systemd service PIDs to the current profile in `gateway status`.

## Problem

`hermes -p <profile> gateway status` was reporting other profiles' PIDs and claiming the named profile's gateway was running, even when it had no gateway.

**Example:**
```
$ hermes -p it-ops gateway status
✓ Gateway is running (PID: 2081267, 2061253)
```

But PIDs 2081267 and 2061253 belong to `default` and `fte-manager` profiles, not `it-ops`.

## Root Cause

The `_get_service_pids` function's systemd branch was always returning PIDs from **ALL** `hermes-gateway*` units, even when `all_profiles=False`. The comment said "systemd always lists every hermes-gateway* unit regardless of scope" — this was the bug.

## Solution

Filter systemd services by the current profile when `all_profiles=False`:

- **Default profile**: Only accept `hermes-gateway.service` (no suffix)
- **Named profiles**: Only accept `hermes-gateway-<profile>.service`

This matches the behavior of the launchd branch (macOS) which already filters by profile.

## Changes

- Modified `hermes_cli/gateway.py`: Added profile filtering to systemd branch in `_get_service_pids`

## Behavior After Fix

**Before:**
```
$ hermes -p it-ops gateway status
✓ Gateway is running (PID: 2081267, 2061253)  # Wrong! These are other profiles
```

**After:**
```
$ hermes -p it-ops gateway status
✗ Gateway is not running  # Correct!
```

## Testing

- ✅ Default profile: only accepts `hermes-gateway.service`
- ✅ Named profiles: only accepts `hermes-gateway-<profile>.service`
- ✅ `all_profiles=True`: still returns all PIDs (for `hermes update`)

---

🤖 Generated with Codebuff

### DIFF
--- a/hermes_cli/gateway.py
+++ b/hermes_cli/gateway.py
@@ -137,8 +137,17 @@ def _get_service_pids(all_profiles: bool = False) -> set:
     pids: set = set()
 
     # --- systemd (Linux): user and system scopes ---
-    # systemd always lists every hermes-gateway* unit regardless of scope.
+    # When all_profiles=False, only include PIDs belonging to the current profile.
     if supports_systemd_services():
+        # Determine the current profile's service name pattern
+        current_profile = None
+        if not all_profiles:
+            try:
+                from hermes_cli.profiles import get_active_profile_name
+                current_profile = get_active_profile_name()
+            except Exception:
+                pass
+        
         for scope_args in [["systemctl", "--user"], ["systemctl"]]:
             try:
                 result = subprocess.run(
@@ -159,6 +168,21 @@ def _get_service_pids(all_profiles: bool = False) -> set:
                     if not parts or not parts[0].endswith(".service"):
                         continue
                     svc = parts[0]
+                    
+                    # When not all_profiles, skip services that don't match current profile
+                    if current_profile and not all_profiles:
+                        # Service name format: hermes-gateway-<profile>.service or hermes-gateway.service (default)
+                        expected_suffix = f"-{current_profile}.service"
+                        is_default_profile = current_profile == "default"
+                        if is_default_profile:
+                            # Default profile: only accept hermes-gateway.service (no suffix)
+                            if svc != "hermes-gateway.service":
+                                continue
+                        else:
+                            # Named profile: only accept hermes-gateway-<profile>.service
+                            if not svc.endswith(expected_suffix):
+                                continue
+                    
                     try:
                         show = subprocess.run(
                             scope_args + ["show", svc, "--property=MainPID", "--value"],


alt-glitch:

Duplicate of #16375: both patches scope systemd gateway PID discovery to the active profile while preserving all-profiles update sweeps. #16375 is the earlier open implementation.

teknium1:
Thanks @wooyongbin3-cpu — correct diagnosis: `_get_service_pids` listed the whole `hermes-gateway*` fleet regardless of scope, so `gateway status` in profile B reported profile A's PID (#97360).

This is now fixed on main by 82d7a13003f (kokhlo, 2026-08-30): `pattern = get_service_name()` when `all_profiles` is false, i.e. the current profile's exact unit, with the fleet glob only for `all_profiles=True`; 9f8557a09f0 updated the docstring and cc9ef823550 kept it through the CLI refactor. For the record, #16375 by @kinsolee (April) was the earliest PR for this symptom. Closing as fixed on main.

PR #97429, opened 2026-08-28
fix(kanban): reject phantom worker reviewers

auroracapital:
## What

- Reject explicit `kanban_request_review(reviewer=...)` values that are not installed Hermes profiles.
- Preserve permissive lower-level `request_review()` behavior for deliberate external or human lanes.
- Treat `review_reopened` as a fresh ready transition for stranded-ready diagnostics.

## Why

A worker can currently park a card on a nonexistent reviewer indefinitely. When self-heal reopens that card, the ready-age diagnostic anchors to task creation and immediately reports a false hours-old critical.

## Regression proof

Before the fix:

- `test_request_review_rejects_missing_profile_without_mutation` fails because the tool moves the card to review.
- `test_stranded_in_ready_uses_review_reopened_as_fresh_ready_transition` fails because the diagnostic reports nine hours ready immediately after reopen.

After the fix, both pass. The suite also proves an installed `verifier` profile works and direct `request_review(... reviewer="external-human-lane")` remains allowed.

## Verification

- `scripts/run_tests.sh tests/tools/test_kanban_tools.py tests/hermes_cli/test_kanban_diagnostics.py tests/hermes_cli/test_kanban_review_lifecycle_complete.py -q -p no:randomly` -> 63 passed
- Ruff on all five changed files -> passed
- `git diff --check` -> passed

Related: #92349 fixes implicit reviewer selection only; this PR covers explicit worker-tool validation and reopen diagnostics.

### DIFF
--- a/hermes_cli/kanban_diagnostics.py
+++ b/hermes_cli/kanban_diagnostics.py
@@ -1003,9 +1003,10 @@ def _rule_stranded_in_ready(task, events, runs, now, cfg) -> list[Diagnostic]:
     # Find the most recent event that put this task into ready.
     # ``created`` covers tasks born ready; ``promoted`` covers parent-
     # done auto-promotion; ``reclaimed`` covers TTL/crash recovery;
-    # ``unblocked`` covers human-driven resumes.
+    # ``unblocked`` covers human-driven resumes; ``review_reopened`` covers
+    # review self-heal reopening a card for a fresh dispatch attempt.
     READY_TRANSITION_KINDS = {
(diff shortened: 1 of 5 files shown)


teknium1:
Merged via #106547 — landed on `main` in 00bcef9b5f. Your tool-layer guard was the right chokepoint; the commit was re-authored from the placeholder `regen <regen@local>` identity to your GitHub noreply so it credits you. The `kanban_diagnostics` hunk was dropped (main already replaced that loop). Thanks @auroracapital.

PR #70801, opened 2026-07-24
fix(desktop): preserve spaces in MEDIA file paths

rille111:
## What does this PR do?

Fixes Hermes Desktop attachments emitted through standalone `MEDIA:` directives when unquoted file paths contain spaces, including paths that arrive across multiple streaming chunks.

The old standalone parser stopped an unquoted value at its first whitespace. A first pass fixed complete lines, but a partial streamed path could still be rendered too early. For example, `.../B17-B` became a permanent link before the later ` Value Extraction and Retirement Receipt.md` chunks arrived.

This revision handles both cases:

- complete standalone paths are bounded by CR/LF rather than whitespace
- ordinary prose such as `MEDIA: the report is ready` remains prose
- pathless standalone and inline `MEDIA:` prose remains unchanged
- consecutive directives anywhere on a line use the next `MEDIA:` marker as an explicit boundary, so each path retains spaces
- the final consecutive or prose-prefixed inline directive remains provisional across streamed chunks
- malformed consecutive lines and whitespace-only directives use linear scans without regex backtracking
- path quotes, whole-directive quotes, and spaced paths remain separate from trailing prose
- standalone `file:` URLs preserve literal spaces and decoded labels
- a trailing unquoted standalone path is marked provisional while streaming; its undecorated source is restored when another text chunk arrives
- provisional state is removed at reasoning, tool, timeline-completion, and final-response boundaries
- chat rendering and the Artifacts view share the same `MEDIA:` extraction logic, without indexing contained partial paths twice
- Windows drive and UNC paths produce standards-compliant encoded `file:` URLs that round-trip through Electron

## Related Issue

No issue number. This revision adds the streamed-path failure reported in #84175 and adapts the typed provisional-source approach from #87807. #85790 also overlaps but includes separate markdown-wrapped `MEDIA:` behavior that remains outside this PR.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `apps/desktop/src/lib/chat-messages/parts.ts`: classify standalone path, inline, and prose intent; preserve complete paths; track streamed provisional path source; clear it on text/reasoning/final boundaries
[...]

PR #96686, opened 2026-08-27
fix(desktop): allow interior spaces in unquoted MEDIA: paths (#96657)

shellybotmoyer:
## Summary

The Desktop renderer's `MEDIA_TAG_RE` and `MEDIA_LINE_RE` regexes in `apps/desktop/src/lib/chat-messages/parts.ts` used `\S+` for unquoted `MEDIA:` paths, which stops at the first space. When the agent delivers a file whose path contains spaces (e.g. `MEDIA:/home/hermes/Morten - Nobly Kickoff - Opening and cue cards EN.docx`), the Desktop app truncates the path at the first space, renders a link to a nonexistent file, and leaves the rest of the filename as literal text.

The messaging platforms (Telegram, WhatsApp, Slack) handle this correctly — `gateway/platforms/base.py:MEDIA_TAG_CLEANUP_RE` uses a tempered-greedy whitespace-tolerant pattern anchored on a known deliverable extension. This PR ports that approach to the TypeScript side.

### Changes

1. **`MEDIA_DELIVERY_EXTS` constant** — mirrors the Python-side extension list so the two surfaces agree on which paths are valid.
2. **`_MEDIA_PATH_ANCHORED` pattern** — replaces the bare `\S+` branch with `(?:~/|/|[A-Za-z]:[/\\])\S+?(?:[^\S\n]+\S+?)*?\.(?:EXT_ALTERNATION)(?=[\s\`"'_,;:)\]}]|MEDIA:|$)`, which:
   - Starts with a path anchor (`~/`, `/`, `X:\` or `X:/`)
   - Allows interior whitespace via tempered-greedy `(?:[^\S\n]+\S+?)*?` pairs
   - Anchors the end on a known deliverable extension
   - Stops at whitespace/punctuation, `MEDIA:`, or end-of-string (prevents runaway absorption into adjacent tags — same #68773 class the Python side guards against)
3. **Both regexes updated** — `MEDIA_LINE_RE` and `MEDIA_TAG_RE` now use `RegExp` constructors with the anchored path branch inserted before the `\S+` fallback.
4. **Test file** — 10 tests covering: simple paths, quoted paths (backtick/double/single), unquoted paths with interior spaces (the #96657 case), Windows paths, `~/`-relative paths, multi-line multiple tags, and non-MEDIA text passthrough.

### Test plan

- [x] `npx vitest run src/lib/chat-messages/parts.test.ts` — 10/10 pass
- [x] `npx vitest run src/lib/chat-messages/` — all pass
[...]

PR #97439, opened 2026-08-28
fix(desktop): allow spaces in MEDIA: paths in Desktop renderer

### LINKED ISSUE #96657: [Bug]: Desktop — `MEDIA:` paths containing spaces are truncated at the first space (messaging platforms extract them correctly)
**Describe the bug**

The `MEDIA:` contract is presented to the model as `MEDIA:/absolute/path/to/file` (`agent/prompt_builder.py:867`, the `desktop` platform prompt) with no instruction to quote. When such a path contains spaces, the Desktop renderer captures only the run up to the first space: the message renders a link to a **nonexistent** file and leaves the rest of the filename behind as literal text.

The messaging platforms handle the identical tag correctly — `gateway/platforms/base.py:2028` (`MEDIA_TAG_CLEANUP_RE`) deliberately allows interior spaces. The Desktop path still uses the naive `\S+` form that base.py's own comment calls *"the old loose ``MEDIA:\\s*\\S+``"*. So the same agent output delivers fine over Telegram/WhatsApp/Slack and breaks in the Desktop app.

**To reproduce**

1. Desktop session against a gateway (we hit it in remote mode; the parsing is client-side either way).
2. Have the agent deliver a file whose path contains spaces:
   ```
   MEDIA:/home/hermes/Morten - Nobly Kickoff - Opening and cue cards EN.docx
   ```
3. The message renders as a link **`Open Morten`**, followed by the literal text `- Nobly Kickoff - Opening and cue cards EN.docx`.
4. Clicking it fails with:
   > Couldn't fetch Morten from the gateway (missing, unreadable, or too large).

**Expected behavior**

The full path is extracted, as it already is on the messaging platforms — link label `Open Morten - Nobly Kickoff - Opening and cue cards EN.docx`, pointing at the real file.

**Root cause**

`apps/desktop/src/lib/chat-messages/parts.ts:13-15`:

```ts
const MEDIA_LINE_RE = /(^|\n)[\t ]*[`"']?MEDIA:\s*(?<line>`[^`\n]+`|"[^"\n]+"|'[^'\n]+'|\S+)[`"']?[\t ]*(\n|$)/g
const MEDIA_TAG_RE  = /[`"']?MEDIA:\s*(?<inline>`[^`\n]+`|"[^"\n]+"|'[^'\n]+'|\S+)[`"']?/g
```

The only space-tolerant branches are the quoted ones; an unquoted path falls through to `\S+`.

For an unquoted spaced path on its own line, `MEDIA_LINE_RE` fails to match **at all** (its trailing `[\t ]*(\n|$)` can't be satisfied once `\S+` stops mid-line), so `renderMediaTags` falls through to `MEDIA_TAG_RE`, which matches mid-line and truncates.

Minimal repro, regexes copied verbatim from `main` (`28ee6ac`):

```js
const MEDIA_LINE_RE = /(^|\n)[\t ]*[`"']?MEDIA:\s*(?<line>`[^`\n]+`|"[^"\n]+"|'[^'\n]+'|\S+)[`"']?[\t ]*(\n|$)/g
const MEDIA_TAG_RE  = /[`"']?MEDIA:\s*(?<inline>`[^`\n]+`|"[^"\n]+"|'[^'\n]+'|\S+)[`"']?/g
// ... unquoteMediaPath / mediaLink / renderMediaTags as in parts.ts

renderMediaTags('MEDIA:/home/hermes/Morten - Nobly Kickoff - Opening and cue cards EN.docx')
// => '[File: Morten](#media:%2Fhome%2Fhermes%2FMorten) - Nobly Kickoff - Opening and cue cards EN.docx'

// quoting is the workaround, and confirms the quoted branch is the only space-safe one:
renderMediaTags('MEDIA:"/home/hermes/Morten - Nobly Kickoff - Opening and cue cards EN.docx"')
// => '[File: Morten - Nobly Kickoff - Opening and cue cards EN.docx](#media:%2Fhome%2Fhermes%2FMorten%20-%20Nobly%20Kickoff%20-%20Opening%20and%20cue%20cards%20EN.docx)'
```

Compare the Python side, `gateway/platforms/base.py:2028`:

```python
(?:~/|/|[A-Za-z]:[/\\])\S+?(?:[^\S\n]+\S+?)*?\.(?:<MEDIA_DELIVERY_EXTS alternation>)
```

`(?:[^\S\n]+\S+?)*?` is exactly the "allow interior spaces, stop at a known deliverable extension" behavior the Desktop side is missing. Verified in Python against the same input, it extracts the full `/home/hermes/Morten - Nobly Kickoff - Opening and cue cards EN.docx`.

**Why this is easy to misdiagnose**

The truncated path then fails at the gateway, and the renderer collapses any fetch failure into the generic notice at `apps/desktop/src/components/assistant-ui/markdown-text.tsx:124` — *"missing, unreadable, or too large"* — which reads as a problem with the file itself. In our session the agent spent several turns re-validating a perfectly good `.docx` and producing an underscore-named copy on the theory that the filename was at fault. (#90580 makes the same point about that message.)

**Suggested fix**

Port the base.py approach into `parts.ts`: allow interior whitespace and anchor the end on a known-deliverable-extension list — ideally sharing one extension list between the Python and TS sides so the two surfaces can't drift again.

A prompt-side "quote paths containing spaces" instruction would also work, but the messaging platforms already accept unquoted spaced paths, so the surfaces should agree rather than the model having to know which one it's talking to.

**Environment**

- Behavior verified against `main` @ `28ee6ac` (regexes copied verbatim into a node repro; Python side executed against the real `MEDIA_TAG_CLEANUP_RE`).
- Originally hit with a Desktop client (macOS) in remote mode against a self-hosted `hermes serve` (`auth_required: true`, `basic` provider), gateway version 0.20.2.

wooyongbin3-cpu:
## Summary

This PR fixes #96657 by updating the Desktop renderer's MEDIA: regex to allow spaces in paths.

## Problem

The Desktop renderer's MEDIA: regex used `\S+` which doesn't allow spaces, causing paths with spaces to be truncated at the first space:

```
MEDIA:/home/hermes/Morten - Nobly Kickoff - Opening and cue cards EN.docx
```

Was rendered as a link to `Morten` instead of the full path.

## Solution

Changed `\S+` to `[^\n]+?` (non-greedy match until end of line) in both `MEDIA_LINE_RE` and `MEDIA_TAG_RE`, allowing interior whitespace in unquoted paths.

This matches the behavior of the Python side (`gateway/platforms/base.py`) which allows interior spaces in MEDIA: paths.

## Changes

- Modified `apps/desktop/src/lib/chat-messages/parts.ts`: Updated regex to allow spaces

---

🤖 Generated with Codebuff

### DIFF
--- a/apps/desktop/src/lib/chat-messages/parts.ts
+++ b/apps/desktop/src/lib/chat-messages/parts.ts
@@ -10,9 +10,9 @@ export function reasoningPart(text: string, timestamp?: number): ChatMessagePart
   return { type: 'reasoning', text, ...(timestamp !== undefined ? { timestamp } : {}) }
 }
 
-const MEDIA_LINE_RE = /(^|\n)[\t ]*[`"']?MEDIA:\s*(?<line>`[^`\n]+`|"[^"\n]+"|'[^'\n]+'|\S+)[`"']?[\t ]*(\n|$)/g
+const MEDIA_LINE_RE = /(^|\n)[\t ]*[`"']?MEDIA:\s*(?<line>`[^`\n]+`|"[^"\n]+"|'[^'\n]+'|[^\n]+?)[`"']?[\t ]*(\n|$)/g
 
-const MEDIA_TAG_RE = /[`"']?MEDIA:\s*(?<inline>`[^`\n]+`|"[^"\n]+"|'[^'\n]+'|\S+)[`"']?/g
+const MEDIA_TAG_RE = /[`"']?MEDIA:\s*(?<inline>`[^`\n]+`|"[^"\n]+"|'[^'\n]+'|[^\n]+?)[`"']?/g
 
 function unquoteMediaPath(value: string): string {
   const trimmed = value.trim()


alt-glitch:

Duplicate of #70801: this fixes the same Desktop unquoted MEDIA-path whitespace parsing defect; #70801 is the broader open parser/resolver implementation.

kvnloo:
exact-head 119dd9be — void if moved

CHECK / pick-one: 1-file `\S+` → `[^\n]+?` in `parts.ts` for #96657; MERGEABLE but naive (can over-swallow trailing prose). Broader/safer open fix is **#96686** (extension-anchored + dedicated tests).

Recommend: close as superseded by #96686.

kvnloo:
exact-head 119dd9be — void if moved

CHECK / pick-one: 1-file `\S+` → `[^\n]+?` in `parts.ts` for #96657; MERGEABLE but naive (can over-swallow trailing prose). Broader/safer open fix is **#96686** (extension-anchored + dedicated tests).

Recommend: close as superseded by #96686.

teknium1:
Resolved on main by #114358 (c48625bf9631af5e93dc15f95fadf36c767afd0c): unquoted MEDIA paths with interior spaces now render as one card in Hermes Desktop, streaming included. Closing as superseded by the landed fix (grammar from #96686, earliest fix #70801); thanks for the report and the PR.

Issue #80680, opened 2026-08-07
[Bug]: Desktop “Show earlier messages” disappears before reaching the true session start


jwtor7:
## Bug description

In Hermes Desktop, older conversation history can become unreachable after switching to a long or compressed session.

Scrolling upward reaches the top of the currently materialized segment rather than loading more history automatically. A **Show earlier messages** control is sometimes present. Clicking it may reveal another segment, but it can disappear before the original beginning of the conversation is reached. Once it disappears, ordinary upward scrolling cannot recover any earlier messages.

This is more than a discoverability problem: the control is driven by the messages already loaded into Desktop, not by whether older durable display history exists.

## Steps to reproduce

1. Use Hermes Desktop with a long session that has undergone context compression.
2. Switch to another session, then switch back to the long session.
3. Scroll to the top of the visible transcript.
4. If **Show earlier messages** appears, click it and continue moving upward.
5. Observe that the control can disappear before the original first message is visible. Further upward scrolling does nothing.

The symptom is intermittent across sessions because short or uncompressed sessions do not cross the same history boundaries.

## Expected behavior

- The transcript should behave like a normal chat history: reaching the top loads the previous page while preserving scroll position.
- Loading should continue until the true beginning of the user-visible conversation is reached.
- A fallback **Show earlier messages** control should remain available while the server reports older display history.
- Switching sessions should not make previously accessible history unreachable.
- Renderer memory limits must remain bounded for tool-heavy transcripts.

## Actual behavior

- Scrolling only traverses the currently materialized tail.
- The button may require repeated clicks, then disappears before the true session head.
- The button state indicates only whether more messages exist in the current client array/window; it does not prove there is no older durable history.
- There is no remaining UI path to request additional history after the control disappears.

## Environment

- Hermes Desktop: `v0.20.0 +21`
- Source/build commit: `01a1037d1e6d7b6eb96a786ef282c3aea4818194`
- Operating system: macOS 26.6
- Python: 3.11.15

## Root-cause analysis

There are two separate bounded windows, both derived from the transcript already loaded into Desktop:

1. `apps/desktop/src/app/chat/index.tsx` resets `windowPages` to one when the assistant runtime/session changes.
2. `apps/desktop/src/app/chat/transcript-window.ts` selects a weighted tail from the in-memory message array.
[...]

Issue #70445, opened 2026-07-24
[Bug]: Desktop remote/VPS — session load is slow, cancels on navigate away, can spin forever, and flashes content then spinner


networthexplained:
### Bug Description

On Hermes Desktop connected to a **remote / VPS** backend, chat session loading is unreliable and slow:

1. **Slow load** — often ~20+ seconds to open a chat.
2. **Load cancels if you leave the chat** — hydration only continues while you stay on that session page; switching away aborts progress; you must babysit the loading screen.
3. **Infinite spinner** — some sessions never finish loading; loading animation runs continuously.
4. **Content flash then spinner** — switching to a chat briefly shows real transcript content, then immediately replaces it with the loading indicator (feels broken / trust-destroying).

These are the same load/hydration pipeline failures under remote latency. **High priority** — remote Desktop is a primary product path.

### Steps to Reproduce

**A — Slow + cancel on leave**
1. Connect Desktop to a remote/VPS Hermes gateway.
2. Click a historical session with substantial history.
3. Note long loading time.
4. While still loading, switch to another session or New session.
5. Return to the original session — load appears restarted / not completed in background.

**B — Infinite load**
1. Open certain sessions (repro intermittent; long history / stalled runtime more common).
2. Observe perpetual session loading state with no recovery UI.

**C — Flash then spinner**
1. Switch to a session that was previously partially loaded or cached.
2. Observe transcript paint briefly, then `loadingSession` UI takes over and content disappears.

### Expected Behavior

- Session history hydrate **continues in background** after navigate away; returning shows progress or completed transcript.
- Reasonable remote timeouts with **retry** (bounded), not silent forever-spinner.
- Never replace already-visible messages with a full-page loader unless hard invalidation is required; prefer inline “Refreshing…” over wiping content.
- Target: interactive transcript well under 20s on typical VPS; progressive render OK.

### Actual Behavior

- ~20s loads common on VPS.
- Leaving the chat stops loading.
- Some chats never load.
- Visible content → spinner regression on switch.

### Affected Component

Desktop session resume / thread hydration / remote gateway (`apps/desktop`)

### Root Cause Hints (code pointers)

| Area | Path |
|------|------|
| Session vs response loading state | `apps/desktop/src/app/chat/thread-loading.ts` — `threadLoadingState(loadingSession, …)` forces `'session'` loader whenever `loadingSession` is true (explains wipe/flash if flag flips true after paint) |
[...]

Issue #69640, opened 2026-07-22
Desktop: session open ignores existing pagination and fetches the full active transcript


ruben2000de:
## Bug description

Desktop's session-read helper always requests the complete active transcript even though the backend already exposes `limit` and `offset` specifically to avoid loading an entire transcript.

`apps/desktop/src/hermes.ts` currently builds:

```ts
path: `/api/sessions/${encodeURIComponent(id)}/messages${suffix}`
```

without pagination parameters. Cold and warm session resume paths then prefetch that full response (`use-session-actions`, session tile delegate, messaging reconciliation, artifacts, export, etc.). On the backend, `/api/sessions/{session_id}/messages` only applies its 500-row clamp when `limit` is supplied; omitted `limit` remains unbounded.

This becomes especially expensive because persisted user/tool messages may contain full `data:image/...;base64,...` values.

## Evidence from a real v0.19.0 database

Read-only aggregate measurements (no transcript content inspected or included):

- 120 sessions / 32,679 message rows; no orphan rows and no session message-count mismatches.
- 69 active message rows contain image data URLs.
- 5 active rows exceed 1 MiB; the largest active row is 7.81 MiB.
- The largest currently active transcript response is about 8.12 MiB for 73 messages.
- Across historical active transcripts, 76.91 MiB of serialized payload is eligible for these unbounded reads.

A single large row is enough to cause a visibly long session-open prefetch and large renderer allocations. The current code comment already recognizes that a "big session" affects resume wall time, but it still requests the whole transcript.

## Reproduction

1. Create a session with a few hundred active messages and several image/tool rows containing base64 image data.
2. Restart Desktop or select the session from a cold state.
3. Observe the request to `/api/sessions/<id>/messages` has no `limit` or cursor.
4. The backend reads, decodes, serializes, transfers, parses, and projects the complete active transcript before the cold prefetch resolves.

## Expected behavior

- Opening a session should fetch a bounded newest-message window.
- Older messages should load on demand as the user scrolls upward.
- Live gateway resume should reconcile against that bounded window rather than requiring a complete REST snapshot.
- Image payloads should ideally be represented by bounded thumbnails/asset references rather than repeated raw data URLs.

## Suggested approach

1. Add reverse/cursor pagination to the endpoint (`before_id`/`before_timestamp`, newest-first window, and `has_more` or total count). Offset-only ascending pagination cannot efficiently request the newest page without first knowing the count.
[...]

PR #97440, opened 2026-08-28
perf(state): bound compacted display tail reads

sashalab:
## What does this PR do?

Desktop requests session messages with `include_compacted=true`. Before this
change, a genuinely compacted session forced `SessionDB.get_messages()` to
fetch, decode, and Python-deduplicate the complete display history before
applying `LIMIT`. In production that turned a 120-message tail request into a
14.7-second stall.

This PR keeps both uncompacted and compacted display reads bounded:

- uncompacted sessions stay on the existing SQL `LIMIT` / `OFFSET` path;
- compacted sessions collapse exact generation copies with SQLite
  `ROW_NUMBER()` before transferring rows to Python;
- only potential composite user carriers are projected in Python, and only
  exact raw rows matching their projected keys are looked up;
- collision detection and page selection share one SQLite read snapshot;
- slow compacted reads log separate collision SQL, normalization, page SQL,
  and decode timings without logging message content.

The existing display-dedupe semantics remain intact for live vs archived
generations, composite carriers, typed display events, tool identities,
`include_inactive`, and pagination.

## Related Issue

Related to #70445, #69640, and #80680.

## Type of Change

- [x] Bug fix
- [x] Performance improvement
- [x] Tests

## How to Test

```bash
scripts/run_tests.sh tests/hermes_state/test_get_messages_include_compacted.py -q
scripts/run_tests.sh tests/hermes_cli/test_web_server.py -k get_session_messages -q
```

Observed locally on macOS / Apple Silicon:

- compacted 3,000-row regression: `3,320 -> 120` content decodes;
- legacy import with 3,000 distinct rows sharing one coarse timestamp:
  `3,120 -> 120` content decodes;
- 30,000 archived rows plus 200 live rows, latest 120 requested: `61.9 ms`
  using SQLite DELETE journal mode.

Verification:

- 17 targeted state tests passed;
- 8 messages-endpoint tests passed;
- 205 Hermes state tests passed; one unrelated live-DB ancestry test fails in
  the restricted local sandbox;
- `ruff check` and `git diff --check` passed;
- an independent review compared 144 combinations of carriers, normal rows,
  pagination, and inactive flags against the previous Python-dedupe behavior.

The complete optional-dependency suite was also attempted in the restricted
desktop sandbox: 38,582 tests passed, while 529 unrelated tests requiring
missing extras, socket binding, process control, or external services failed.

## Checklist

- [x] Conventional commits
- [x] Targeted regression tests
- [x] macOS / Apple Silicon verification
- [x] No message contents added to diagnostics
- [ ] Complete all-extras suite green in upstream CI

### DIFF
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -12014,7 +12014,7 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
         else:
             active_clause = " AND active = 1"
         needs_display_dedupe = include_compacted
-        if include_compacted:
+        if include_compacted and not include_inactive:
             # Desktop always opts into compacted display history, including
             # for sessions that have never compacted. Avoid turning those
             # ordinary bounded reads into a full transcript materialization:
@@ -12030,7 +12030,7 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
                     ).fetchone()
                     is not None
                 )
-            if not needs_display_dedupe and not include_inactive:
+            if not needs_display_dedupe:
                 # Keep the fast query active-only even if a concurrent
                 # compaction commits after the probe. That request may see the
                 # new compacted tail on its next refresh, but it cannot mix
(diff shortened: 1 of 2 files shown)


kshitijk4poor:
**Triage finding (perf/P2 sweep) — measured regression on the compacted path; please don't merge as-is.**

Cherry-picked onto `origin/main` 1cb3ab6173 (two conflict resolutions — the uncompacted fast-path probe from your first commit is already on main). 17/17 tests pass, but none of them times anything, so this went unnoticed:

Fixture: one session with 6 compaction generations × 3000 archived rows (`active=0, compacted=1`) + 300 live rows; `get_messages(sid, limit=120, latest=True, include_compacted=True)`, median of 5:

| row content size | main | this PR |
|---|---|---|
| 300 B | 95 ms | **244 ms** (2.6× slower) |
| 5 KB | 162 ms | **2175 ms** (13× slower) |
| 20 KB | 345 ms | **8877 ms** (25× slower) |

(Also with byte-identical copies across generations, i.e. the real compaction shape: 300 B 93 → 223 ms; 5 KB 184 → 1706 ms.) The uncompacted path is unchanged at 0.6 ms.

Cause: `ROW_NUMBER() OVER (PARTITION BY role, content, timestamp, tool_call_id, tool_calls, tool_name …)` makes SQLite sort the **full `content` blobs of every display row** into a temp B-tree (`EXPLAIN QUERY PLAN` → `USE TEMP B-TREE FOR ORDER BY`), and the CTE runs twice (collision query + page query). That is strictly more work than main's single index scan + Python dedupe; the 14.7 s stall you saw is plausible, but this moves the cost rather than bounding it, and it gets worse exactly when rows are large.

Fix direction that would keep the intent: partition on a cheap key — `(timestamp, role, length(content), tool_call_id, tool_name)` or a stored content hash column — and dedupe only the resulting small candidate set in Python; keep the `needs_display_dedupe` fast path (already on main). A timing assertion in the test (e.g. the 5 KB fixture above under a generous bound) would pin it.

Bench: `bench/s2_state_reads.py` (fixture A, `ROWBYTES` env for row size).

kshitijk4poor:
Closing this one — not for the idea, which is right (the compacted display tail should not re-read the whole archived history), but because the implementation as written is a measured regression on the path it targets, and the maintainer decision is to not carry it as an open item.

Numbers from the triage bench (`get_messages(include_compacted=True, latest, limit=120)` on a session with 18k archived + 300 live rows, fresh `origin/main` vs this PR cherry-picked, macOS arm64, medians of 5):

| row size | main | this PR |
|---|---|---|
| 300 B | 95 ms | 244 ms (2.6× slower) |
| 5 KB | 162 ms | 2175 ms (13×) |
| 20 KB | 345 ms | 8877 ms (25×) |

Root cause: the dedupe uses `ROW_NUMBER() OVER (PARTITION BY content ...)`, which forces SQLite to sort the full `content` blobs into a temp B-tree before the window function runs — the cost scales with total archived bytes, not with the 120-row tail. The PR's own tests pass because none of them measure time.

If you want to take another run at it, the shape that avoids this is to dedupe on a cheap key — `rowid`/`id` (or a stored content hash) — and select the tail by id range first, then hydrate content only for the ≤120 rows kept. A perf test asserting the query touches O(limit) rows (or a wall-clock bound on a large seeded DB) would keep it from regressing again. Happy to review a follow-up along those lines; the bench script is available if it helps.

Thank you for the work on this, @sashalab — the problem you identified is real.

PR #88097, opened 2026-08-17
fix(cli): stop Shift+Space CSI bytes leaking into buffer (#88071)

webtecnica:
## What Changed

Adds `install_keypress_data_normalization()` in `hermes_cli/pt_input_extras.py`, called from CLI startup alongside the other alias installers. It patches `Vt100Parser._call_handler` so that when an extended-key sequence maps to a single plain character (Shift+Space → `' '`, Shift+letter → the uppercase letter, keypad digits/operators), the `KeyPress.data` field carries the **mapped character** instead of the raw CSI bytes.

The patch is idempotent (sentinel flag) and defensive (import failures degrade to a no-op), matching the existing `Vt100Parser.feed` bracketed-paste patch pattern.

## Root Cause

`Vt100Parser._call_handler` builds `KeyPress(key, match.group(0))` — the **key** was correctly remapped by the alias table (#86866 follow-up, #87511/#87630), but the **data** field still held the full raw prefix (e.g. `\x1b[32;2u`). prompt_toolkit's default character-insert binding (`self-insert`, `basic.py`) inserts `event.data`, so the raw CSI bytes landed in the buffer:

- `Shift+Space xterm` → `'ab\x1b[27;2;32~cd'`
- `Shift+Space kitty` → `'ab\x1b[32;2ucd'`

For a plain space both fields are `' '`, which is why the bug was invisible there. The same leak class affected **every** string-valued alias: Shift+letter, keypad digits, keypad operators. The previous test only asserted on `key_press.key == ' '` (parser output), so it passed while the buffer-level behavior leaked.

## Verification

- New parser-level tests assert `KeyPress.data` equals the mapped character for Shift+Space (both xterm modifyOtherKeys and kitty CSI-u formats), Shift+letter, keypad digits, and that plain space typing is unchanged.
- New end-to-end test feeds the sequences through a real `Application` + `Buffer` (the repro from the issue) and asserts the buffer contains `'ab cd'` with no raw CSI bytes.
- `tests/cli/test_modify_other_keys_aliases.py`: 178 passed.
- `tests/cli/test_cli_shift_enter_newline.py`, `test_cli_cmd_backspace.py`, `test_cli_terminal_shortcuts.py`: 20 passed.

PR #97443, opened 2026-08-28
fix(cli): deliver literal-char key mappings with the char as KeyPress.data

### LINKED ISSUE #97413: Shift+letter inserts raw escape sequence instead of uppercase char in interactive chat CLI (Ghostty, regression)
## Bug

In the interactive chat CLI (`hermes` with no subcommand), pressing
Shift+<letter> inserts the raw escape sequence into the prompt instead
of the uppercase character. Example — Shift+S produces:

    ^[[27;2;83~

instead of `S`.

## Environment

- Terminal: Ghostty (`TERM_PROGRAM=ghostty`, `TERM=xterm-ghostty`)
- Connection: direct SSH (no mosh, no tmux/screen)
- macOS, Python 3.11.16
- hermes-agent installed via fresh `git clone` of `main`

## Regression

A previous, older installation of hermes-agent (same machine, same
Ghostty config) did not have this problem. This appears specific to
the current `main`.

## What I found tracing it

`_is_ghostty_terminal()` / `_enable_extended_enter_keys()` in `cli.py`
detect Ghostty and push xterm `modifyOtherKeys` level 2 for it
specifically (the "Ghostty exception" path, only modifyOtherKeys, no
Kitty CSI-u).

`install_modify_other_keys_aliases()` in `hermes_cli/pt_input_extras.py`
is supposed to decode `\x1b[27;2;83~` back to `'S'` via
`ANSI_SEQUENCES`. Tested this in isolation (calling the installer
directly and checking the table):

    seq in table: True
    mapped to: S

So the mapping is correctly installed when tested standalone, but the
live interactive session still shows the raw sequence — suggesting
either a timing/ordering issue at real startup, or the live
prompt_toolkit input parser isn't picking up the patched table.

## Workaround

Forcing `TERM_PROGRAM=` and `TERM=xterm-256color` when launching
`hermes` avoids triggering the Ghostty extended-keyboard path entirely
and restores correct behavior:

    TERM_PROGRAM= TERM=xterm-256color hermes

liuhao1024:
## What does this PR do?

Fixes the Ghostty regression where pressing Shift+<letter> in the interactive chat CLI inserted the raw escape sequence (`^[[27;2;83~`) instead of the uppercase character (#97413).

Root cause is one layer deeper than the alias table: `install_modify_other_keys_aliases()` maps `ESC[27;2;<cp>~` to a **bare uppercase string** in `ANSI_SEQUENCES`. prompt_toolkit's `Vt100Parser._call_handler(match, prefix)` then emits `KeyPress(key='S', data='\x1b[27;2;83~')` — the matched prefix is always the raw sequence — and the stock `Keys.Any` binding inserts `event.data` into the buffer. So the table lookup succeeds (the key *is* 'S', which is why the existing key-only tests pass) while the visible insert is the 8-byte raw sequence. Stock prompt_toolkit never maps escape sequences to bare characters, so nothing upstream corrects the data.

Ghostty is the terminal that actually triggers this: the Ghostty exception path pushes only modifyOtherKeys level 2 (no Kitty CSI-u), and modifyOtherKeys=2 re-encodes every Shift+letter as `ESC[27;2;<cp>~`. Kitty-protocol terminals keep sending the plain byte for printable Shift+letter, which is why the CSI-u side of the same mapping never leaked.

The fix wraps `Vt100Parser._call_handler` (idempotent, module sentinel — same pattern as the existing bracketed-paste patch) and substitutes the character as `data` for Hermes-installed literal-char entries: single non-`Keys` character keys whose `data` is a different escape sequence. This also repairs the sibling literal mappings in one pass: Shift+Space (`ESC[27;2;32~` → `" "`) and kitty keypad digits (`ESC[57399u` → `"0"`). Plain typed characters (`key == data`), `Keys`-enum targets (Ctrl combos keep the stock data semantics), and tuple expansions (`(Escape, x)` — data is `''` after the first element) all fall outside the condition and are untouched.

## Related Issue

Fixes #97413

## Type of Change

## Changes Made

- `hermes_cli/pt_input_extras.py`: added `_install_literal_char_data_patch()` — wraps `Vt100Parser._call_handler` so literal-char table entries carry the char as `KeyPress.data`; called at the end of `install_modify_other_keys_aliases()` next to the existing prefix-cache clear.
- `tests/cli/test_modify_other_keys_aliases.py`: new regression tests asserting the **data** layer (the existing suite only asserted `kp.key`): Shift+letter (both modifyOtherKeys and CSI-u forms, all 26 letters), Shift+Space, kitty keypad digit, plus two guards — plain typed chars keep `key == data`, and Ctrl combos keep the raw sequence as data.

## How to Test

1. `pytest tests/cli/test_modify_other_keys_aliases.py tests/cli/test_cli_shift_enter_newline.py tests/cli/test_cli_cmd_backspace.py tests/cli/test_cli_terminal_shortcuts.py tests/hermes_cli/test_curses_arrow_keys.py -q` — 285 passed. Full `pytest tests/cli/ -q`: 1340 passed; the single `test_resume_quiet_stderr` failure is a pre-existing order-coupled flake reproduced identically on clean upstream/main (passes when run alone, with or without this diff).
2. End-to-end repro on macOS via pty (`TERM_PROGRAM=ghostty TERM=xterm-ghostty hermes --yolo`): before the fix, writing `\x1b[27;2;83~` to the pty rendered `^[[27;2;83~` (ESC shown as the blue control-char form) in the composer; after the fix the same write inserts a clean `S` (observed result: composer renders `S`; `\x1b[27;2;32~` inserts a space).
3. A `sitecustomize` probe on the unpatched build confirmed the failure chain directly: `_call_handler` received `key='S'` with the raw sequence as data, i.e. the parser layer was correct and only the data payload was wrong.

### Code

- [x] I've tested on my platform: macOS 15 (arm64), pty-driven Ghostty emulation

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (the new function is fully documented in its docstring)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (parser-level patch applies identically on all platforms; no Windows-specific input path touches these sequences)

### DIFF
--- a/hermes_cli/pt_input_extras.py
+++ b/hermes_cli/pt_input_extras.py
@@ -48,6 +48,53 @@ def _clear_vt100_prefix_cache() -> None:
         pass
 
 
+def _install_literal_char_data_patch() -> bool:
+    """Deliver literal-char table entries with the char itself as ``KeyPress.data``.
+
+    ``ANSI_SEQUENCES`` entries installed for Shift+letter (→ uppercase char,
+    e.g. ``\\x1b[27;2;83~`` → ``'S'``), Shift+Space (→ ``' '``) and kitty
+    keypad digits map an escape sequence to a *bare character*. prompt_toolkit
+    then emits ``KeyPress(key=<char>, data=<raw escape sequence>)`` because
+    the parser passes the matched prefix as ``data`` — and the stock
+    ``Keys.Any`` binding inserts ``event.data``, so the raw sequence leaks
+    into the prompt buffer as literal text (#97413: Ghostty under
+    modifyOtherKeys=2 re-encodes every Shift+letter this way).
+
+    Stock prompt_toolkit never maps escape sequences to bare characters, so
+    wrapping ``Vt100Parser._call_handler`` to substitute the char as ``data``
+    only affects Hermes-installed literal entries: plain typed characters
+    (``key == data``), ``Keys`` enum targets and tuple expansions
+    (``data=''`` after the first element) all fall outside the condition.
+    Idempotent via a module sentinel.
+    """
+    try:
+        import prompt_toolkit.input.vt100_parser as _vt100_mod
+        from prompt_toolkit.keys import Keys
+    except Exception:
+        return False
+
+    if getattr(_vt100_mod, "_hermes_literal_char_patched", False):
+        return True
+
+    _orig_call_handler = _vt100_mod.Vt100Parser._call_handler
+
+    def _patched_call_handler(self_parser, key, insert_text):
+        if (
+            isinstance(key, str)
+            and not isinstance(key, Keys)
+            and len(key) == 1
+            and isinstance(insert_text, str)
+            and insert_text.startswith("\x1b")
+            and insert_text != key
+        ):
+            insert_text = key
+        return _orig_call_handler(self_parser, key, insert_text)
+
+    _vt100_mod.Vt100Parser._call_handler = _patched_call_handler
+    _vt100_mod._hermes_literal_char_patched = True
+    return True
+
+
 def install_shift_enter_alias() -> int:
     """Map Shift+Enter byte sequences to the (Escape, ControlM) key tuple
     that Alt+Enter produces, so the existing Alt+Enter newline handler
@@ -491,6 +538,11 @@ def install_modify_other_keys_aliases() -> int:
     # created before this install (or in earlier tests) can't misparse.
     if changed:
         _clear_vt100_prefix_cache()
+
+    # Literal-char entries installed above (Shift+letter, Shift+Space, kitty
+    # keypad digits) must carry the char as KeyPress.data, or the raw escape
+    # sequence leaks into the prompt buffer via the Keys.Any insert (#97413).
+    _install_literal_char_data_patch()
 
     return changed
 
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #88097: both normalize Vt100Parser KeyPress.data for character-valued ANSI aliases so raw CSI sequences are not inserted. #88097 is the earlier, broader open implementation.

liuhao1024:
Agreed with the duplicate assessment — closing in favor of #88097.

I independently reviewed #88097's diff against this one: it patches the same layer (`Vt100Parser._call_handler`) for the same root cause (the mapped single-char key keeps the raw CSI sequence as `KeyPress.data`, and `self-insert` inserts `event.data`), and its coverage matches everything this PR fixes — Shift+letter in both xterm modifyOtherKeys (`ESC[27;2;<cp>~`) and kitty CSI-u (`ESC[<cp>;2u`) forms, Shift+Space, and keypad digits — plus a buffer-level end-to-end test this PR didn't have. Earlier and broader; no reason to race it.

One piece of context worth carrying over: #97413 (the Ghostty report this PR fixes) is the same root cause from the terminal side — Ghostty's exception path pushes only modifyOtherKeys level 2 (no Kitty CSI-u), and modifyOtherKeys=2 re-encodes printable Shift+letter as `ESC[27;2;<cp>~`, which is exactly the sequence class #88097 already normalizes and tests (`\x1b[27;2;97~` for Shift+a). So #88097 should close #97413 as well once merged; I've noted this on the issue.

Thanks @webtecnica for the earlier implementation.

PR #97444, opened 2026-08-28
fix(desktop): use stored session identity for reactions

### LINKED ISSUE #97277: [Bug] Desktop: "Could not react — No active session" when $activeSessionId is unset (guard fires before any RPC)
## Summary

In the **Hermes Desktop** app, attempting to react to a message can fail with a toast **"Could not react — No active session"**, even when a conversation is visibly open and on screen. This is a *different* failure from the `4040 "message not found"` class tracked in #80670: it fires **before any RPC is sent**, in the client-side guard, when the renderer's active-session id is falsy.

## Repro

1. Open Hermes Desktop (Electron) on a conversation/session.
2. Tap a reaction (tapback) on a message.
3. Observe toast: `Could not react — No active session`.

Observed on **Windows 11**, Hermes Agent v0.20.6 (2026.8.27), upstream `7b5e1911`. A screenshot of the toast is attached in an internal capture (`composer-images\composer_2026-08-28_16-08-21-164_65afae.png`); the toast text is verbatim as reproduced here.

## Root cause (client-side guard, fires before RPC)

`apps/desktop/src/store/reactions.ts` → `toggleMessageReaction`:

```ts
const sessionId = $activeSessionId.get()      // line 61
const gateway = activeGateway()               // line 62

if (!sessionId || !gateway) {
  notifyError(
    new Error(!sessionId ? 'No active session' : 'Gateway not connected'),
    'Could not react'
  )                                           // lines 64-68
  return
}
```

`$activeSessionId` is declared in `store/session.ts` as `atom<string | null>(null)` and is only ever written through `setActiveSessionId`. `toggleMessageReaction` reads it **directly with no fallback**: if the atom is `null` at the moment the user taps a reaction, the toast fires and no `message.react` RPC is ever attempted. The reaction is a no-op on a visibly live conversation.

This is a renderer session-binding / state-loss path: the conversation and its reaction affordances render, but the `$activeSessionId` atom is unset (or was cleared) — e.g. after a context/profile switch, viewing a historical/resumed session, or a race where the message list paints before the active-session binding is set.

## Why this is distinct from #80670 / not covered by #95833

- **#80670** tracks the server-side `message.react` handler returning `4040 "message not found in this session"` — the session *is* bound but the durable row id / session key can't be resolved (`tui_gateway/methods_session.py`, `SessionDB.set_message_reaction` returning `None`).
- **#95833** (open fix for the #80670 class) routes reactions through `requestForOwnedSession` and resolves the write key as `session_key or session_id`. That work all happens *after* the client passes the `!sessionId` guard.

This bug is upstream of both: it fires at `reactions.ts:64-68` when `$activeSessionId` is falsy, so it short-circuits **before** the RPC that #95833 fixes. A fix for the 4040/routing class will not touch this path — if the atom is `null`, the user still sees "No active session" regardless of routing.

## Suggested direction

- Give the reaction path a bound-session fallback rather than a hard `!sessionId` early-return — resolve the owning session the same way other session-scoped actions do (e.g. the tile/owner-session ladder already used elsewhere in the desktop shell), or derive the session id from the message the user is reacting to.
- Treat "active session id unset but a message is mounted" as a diagnosable state rather than a generic false-negative: log which shape fired (null activeSessionId vs missing gateway) so it doesn't masquerade as a user error or as the 4040 class.
- The two branches already coexist in one guard (`No active session` / `Gateway not connected`) — splitting them into distinct, actionable errors (or recovery affordances) would make the state failure self-explanatory.

## Environment

- Hermes Agent v0.20.6 (2026.8.27), upstream `7b5e1911`
- Hermes Desktop (Electron, `apps/desktop/*`)
- OS: Windows 11
- First observed 2026-08-28

Related: #80670 (4040 "message not found" class), #95833 (open fix for that class).

UniversePeak:
## What does this PR do?

Fixes the renderer-side failure in #97277 where the desktop reaction path shows `Could not react — No active session` before sending any RPC when a visible stored conversation has no runtime `$activeSessionId` during resume or context re-binding.

## Related Issue

Fixes #97277

## Type of Change

## Changes Made

- Resolve the reaction session identity from `$activeSessionId`, falling back to `$selectedStoredSessionId` for mounted stored conversations.
- Pass the session identity from `SessionView` into picker and double-click reaction actions, including tile views.
- Route `message.react` through `requestForOwnedSession` so the write follows the session owner rather than the ambient gateway.
- Preserve the no-session guard for genuinely new drafts.
- Add focused regression coverage for stored-session fallback and draft rejection.

## How to Test

1. `npm test -- --project ui src/store/reactions.test.ts`
2. `npm run typecheck`
3. `npm run lint -- --quiet`
4. `npx prettier --check src/store/reactions.ts src/store/reactions.test.ts src/components/assistant-ui/thread/use-message-reactions.ts`

All focused tests, typecheck, lint, and formatting checks pass.

## Checklist

- [x] I've considered cross-platform impact (Windows, macOS) per the compatibility guide
- [x] I've updated relevant documentation — N/A
- [x] I've updated config documentation — N/A

## Screenshots / Logs

N/A — renderer state/routing fix covered by focused tests.

### DIFF
--- a/apps/desktop/src/components/assistant-ui/thread/use-message-reactions.ts
+++ b/apps/desktop/src/components/assistant-ui/thread/use-message-reactions.ts
@@ -2,6 +2,7 @@ import { useAuiState, useMessageRuntime } from '@assistant-ui/react'
 import { useStore } from '@nanostores/react'
 import { type MouseEvent, useCallback } from 'react'
 
+import { useSessionView } from '@/app/chat/session-view'
 import type { ChatMessage } from '@/lib/chat-messages'
 import { triggerHaptic } from '@/lib/haptics'
 import { QUICK_REACTIONS, toggleMessageReaction } from '@/store/reactions'
@@ -41,12 +42,13 @@ function commitReaction(
   role: ChatMessage['role'],
   rowId: number | undefined,
   reactions: MessageReaction[],
-  emoji: null | string
+  emoji: null | string,
+  sessionId: null | string
 ): void {
   // Flip the UI immediately — a tapback is direct manipulation and must never
   // wait on a round-trip. Persistence follows in the background.
   setLocalReaction(messageId, emoji)
-  void toggleMessageReaction({ id: messageId, role, rowId, reactions } as ChatMessage, emoji)
+  void toggleMessageReaction({ id: messageId, role, rowId, reactions } as ChatMessage, emoji, 'user', sessionId)
 }
 
 /**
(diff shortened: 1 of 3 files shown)


UniversePeak:
This is still ready for review: the stored-session fallback keeps the no-session guard for new drafts, and the current head is clean/rebaseable with the required checks passing. It has no review comments or requested changes so far. Could someone take a look when available?

Issue #97427, opened 2026-08-28
[Bug]: GPT-5.6 Responses rejects replayed assistant message when Hermes strips its required reasoning item ID


salehelsayed:
### Bug Description

 When using gpt-5.6-sol through a custom OpenAI-compatible Responses endpoint (api_mode: codex_responses, store:false, reasoning enabled), Hermes eventually fails an automatic post-tool continuation with:

  Item 'msg_...' of type 'message' was provided without its required 'reasoning'
  item: 'rs_...'

  Hermes correctly persists both the rs_* reasoning item and linked msg_* assistant message. During replay, however, it removes the reasoning item's id while retaining the assistant message's id.

  Stored:
    reasoning{id: rs_..., encrypted_content} -> message{id: msg_...}

  Sent:
    reasoning{encrypted_content}             -> message{id: msg_...}

  The API therefore sees a native msg_* item whose required rs_* identity is missing. This reproduced across four fresh sessions with different item IDs, so it is not a single corrupt session, compression, or stale state.

### Steps to Reproduce

 1. Configure a custom OpenAI-compatible endpoint:

     model:
       default: gpt-5.6-sol
       provider: custom:<redacted>
       base_url: http://localhost:<port>/openai/v1
       api_mode: codex_responses

     agent:
       reasoning_effort: medium

  2. Start a fresh Hermes CLI or Desktop conversation.

  3. Request multiple sequential tool calls, for example:

     "Use tools sequentially. First run `pwd`. After seeing that result, run `ls`.
     Do not combine the commands. Then summarize the results."

  4. A Responses turn emits a reasoning item, sometimes a phase=commentary
     message, and a function call.

  5. After the tool result, Hermes replays the output for the next continuation.

  6. The continuation fails with HTTP 400 because the msg_* item retains its ID
     while the preceding reasoning item no longer has its rs_* ID.

### Expected Behavior

Hermes should preserve a valid Responses item graph across tool continuations. It must either replay both linked rs_* and msg_* identities, or strip both and reconstruct the history as plain role/content messages.

### Actual Behavior

  HTTP 400 invalid_request_error:

  Item 'msg_<redacted>' of type 'message' was provided without its required
  'reasoning' item: 'rs_<redacted>'.

Repeating the request fails deterministically even though Hermes classifies the provider error as retryable.

### Affected Component

Agent Core (conversation loop, context compression, memory), CLI (interactive chat)

### Messaging Platform (if gateway-related)

[...]

PR #97445, opened 2026-08-28
fix(codex): avoid orphaned Responses message IDs

### LINKED ISSUE #97442: [Bug]: GPT-5.6 Responses rejects replayed message after Hermes strips its reasoning ID
### Bug description

With `api_mode: codex_responses` and `store: false`, a follow-up GPT-5.6 request can fail when Hermes replays a native assistant message item but removes the reasoning-item identity that the message depends on.

The upstream error is:

```text
Item 'msg_…' of type 'message' was provided without its required 'reasoning' item: 'rs_…'.
```

Hermes persists both output items, but the outgoing continuation currently has this asymmetric shape:

```json
[
  {"type": "reasoning", "encrypted_content": "…", "summary": []},
  {"type": "message", "id": "msg_…", "role": "assistant", "phase": "final_answer", "status": "completed", "content": [{"type": "output_text", "text": "…"}]}
]
```

The `rs_…` ID is absent while the dependent `msg_…` ID remains.

### Steps to reproduce

1. Configure a Responses-compatible provider with `api_mode: codex_responses`.
2. Use a GPT-5.6-family model with reasoning enabled.
3. Complete a turn whose `response.output` includes both a reasoning item and a native assistant message item.
4. Send a follow-up message so Hermes manually replays the prior output.
5. Observe HTTP 400 before inference.

This reproduced deterministically across fresh sessions. The persisted session state contained both original IDs; the adapter removed only the reasoning ID while building the next request.

### Expected behavior

Hermes should never emit a provider-issued output-item ID unless the identities of all prerequisite native items are also present and compatible. If reasoning IDs must be removed for `store:false` compatibility, linked assistant message IDs should be removed too while preserving content, status, and phase.

### Actual behavior

Hermes strips `reasoning.id` in both conversion and preflight, but preserves short non-GitHub `message.id` values. GPT-5.6 rejects the resulting partial native item graph.

### Root cause

This is a latent interaction between:

- #10217, which intentionally stripped reasoning IDs because some GPT-5.4-compatible proxies returned 404 with `store:false`.
- #15782, which later added exact assistant-message replay for cache affinity and retained `message.id`.

Each behavior was tested separately, but no regression combined a linked reasoning/message pair.

Related: #63257 and #67667.

OpenAI's stateless reasoning guidance says to preserve the complete output history; its phase guidance also supports replaying assistant content and phase without a native item ID:

- https://developers.openai.com/api/docs/guides/reasoning#preserve-reasoning-without-stored-responses
- https://developers.openai.com/api/docs/guides/reasoning#phase-parameter

### Environment

- Hermes 0.20.6 (also reproduced against current source behavior)
- Python 3.11.15
- OpenAI Python SDK 2.24.0
- macOS
- Provider type: custom Responses-compatible proxy

salehelsayed:
## What does this PR do?

Fixes #97442.
Related to #97427, which appears to report the same GPT-5.6 Responses replay failure.

Hermes intentionally strips replayed reasoning-item IDs when `store=false` because some Responses-compatible endpoints otherwise try to look them up server-side. A dependent native assistant `msg_*` ID can nevertheless survive, producing an invalid partial item graph such as:

```text
reasoning{encrypted_content} -> message{id: msg_*}
```

GPT-5.6 rejects that request because the message identity depends on the stripped `rs_*` reasoning identity.

This refreshed implementation is rebased onto current `main` and enforces the invariant at the final provider-call boundary, after request middleware and Responses preflight:

- if a native `reasoning` item starts an output group, dependent typed assistant `message` IDs are removed until a real group boundary;
- a `function_call` does not prematurely close the reasoning group;
- `function_call_output` and ordinary role messages close the group;
- if encrypted reasoning replay has been disabled for recovery, all typed assistant message IDs are omitted because their original dependency can no longer be proven;
- independent message IDs are preserved, so the existing cache optimization remains where it is demonstrably safe;
- message content, status, phase, and tool-call pairing are unchanged.

## Current-main refresh

The original PR head had become conflicted after substantial upstream refactoring. On 2026-09-11 the branch was rebuilt as a single focused commit on current `main` (`ad03f20dd61919ca2135d6904e787a94284aacaf`) rather than mechanically resolving the old diff.

Current changes:

- `agent/responses_replay_identity.py`
  - Adds the dependency-closure guard for Responses replay identities.
- `agent/turn_api_call.py`
  - Applies the guard at the final provider-call boundary after middleware and preflight.
- `tests/agent/test_responses_replay_identity.py`
  - Covers linked reasoning/message replay, function-call grouping, tool-output boundaries, independent message IDs, recovery mode, and non-mutation of the original request.
- `contributors/emails/saleh.fekry@gmail.com`
  - Adds the standard contributor mapping for `salehelsayed`.

## Type of Change

- [x] 🐛 Bug fix (non-breaking)
- [x] ✅ Regression tests

## Validation

The refreshed head is conflict-free and GitHub reports it as mergeable against current `main`.

Hosted CI, Docker, and Nix workflows were triggered for the refreshed head but currently show `action_required`, so they require maintainer approval before they can execute for this external-contributor PR.

## Checklist

- [x] Based on current upstream `main`
- [x] Focused only on the Responses replay identity bug
- [x] Preserves independent `msg_*` IDs
- [x] Adds focused regression coverage
- [x] Does not change architecture, configuration, or user-facing behavior outside this fix
- [ ] Hosted CI completed on the refreshed head (awaiting maintainer workflow approval)

### DIFF
--- a/agent/responses_replay_identity.py
+++ b/agent/responses_replay_identity.py
@@ -1,0 +1,77 @@
+"""Responses replay identity safety.
+
+Provider-issued assistant ``message`` IDs can depend on a preceding native
+``reasoning`` item identity. Hermes intentionally strips reasoning IDs when
+replaying stateless Responses history (``store=False``), so a dependent
+``msg_*`` ID must not survive on the outgoing request by itself.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+
+def enforce_responses_replay_identity_closure(
+    api_kwargs: Any, *, drop_all_message_ids: bool = False,
+) -> Any:
+    """Return request kwargs with unsafe Responses ``message.id`` values removed.
+
+    The function is deliberately a final wire guard:
+
+    * after a ``reasoning`` item, typed assistant ``message`` IDs are removed
+      until a real output-group boundary is reached;
+    * a ``function_call`` does not close the group because commentary/final
+      messages may still belong to the same reasoning output group;
+    * ``function_call_output`` and ordinary role messages close the group;
+    * when reasoning replay has been disabled for the session, all typed
+      assistant message IDs are removed because their original dependency can
+      no longer be proven from the remaining history.
+
+    Only the native identity is removed. Content, status, phase, tool pairing,
+    and independent message IDs are preserved. The input object is not mutated.
+    """
+    if not isinstance(api_kwargs, dict):
+        return api_kwargs
+    raw_items = api_kwargs.get("input")
+    if not isinstance(raw_items, list):
+        return api_kwargs
+
+    pending_reasoning_identity = False
+    changed = False
+    normalized_items: list[Any] = []
+
+    for raw_item in raw_items:
+        if not isinstance(raw_item, dict):
+            normalized_items.append(raw_item)
+            continue
+
+        item_type = raw_item.get("type")
+        if item_type == "reasoning":
+            # Preflight intentionally removes reasoning.id for store=False.
+            # Any native message identity in this output group must degrade
+            # symmetrically or the server sees an orphaned msg_* item.
+            pending_reasoning_identity = True
+
+        item = raw_item
+        if item_type == "message" and (
+            pending_reasoning_identity or drop_all_message_ids
+        ) and "id" in raw_item:
+            item = dict(raw_item)
+            item.pop("id", None)
+            changed = True
+
+        normalized_items.append(item)
+
+        # Tool execution completes the reasoning/tool output group. Untyped
+        # role messages are also explicit conversation boundaries. A bare
+        # function_call is intentionally NOT a boundary.
+        if item_type == "function_call_output" or (
+            item_type is None and raw_item.get("role") in {"user", "assistant"}
+        ):
+            pending_reasoning_identity = False
+
+    if not changed:
+        return api_kwargs
+    sanitized = dict(api_kwargs)
+    sanitized["input"] = normalized_items
+    return sanitized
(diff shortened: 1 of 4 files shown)


ehz0ah (commented):
Reviewed exact head `e3a6ba38287c`. The change restores the Responses replay invariant without reintroducing store-false reasoning ID lookups: linked `msg_*` IDs are removed whenever their `rs_*` identity cannot be replayed, while message content, phase, status, and tool-call pairing remain intact. The same rule is applied to conversion, preflight, normalization, stale-replay pruning, and invalid-encrypted-content recovery.

I reproduced the invalid graph on current `main` and confirmed that this head produces the dependency-closed graph. I also ran the focused 10-file suite through `scripts/run_tests.sh`: 672 passed. Ruff, compileall, and `git diff --check` passed. I found no correctness blockers in the submitted implementation. The branch currently conflicts with `main`, so it will need a rebase before merge.

teknium1:
Thanks @salehelsayed — the work in this PR has landed on `main` via:

- #115733 — fix(codex): GPT-5.6 Responses continuations no longer 400 on a replayed message whose reasoning id was stripped (#97427, #97442, salvage #97445) (landed in `4d1d3d05a358`)

Your contribution is credited there (cherry-picked authorship / co-author trailer or credit in the PR body; see the linked PR for what was kept and what was trimmed). Closing this one as landed / superseded so the backlog reflects reality. If something in your original diff is still missing on current `main`, please comment and we'll reopen or follow up.

PR #97450, opened 2026-08-28
fix(update): only require fleet verification for gateway runtimes

### LINKED ISSUE #97332: [Bug]: dashboard-only runtime makes successful `hermes update` exit 1 on empty gateway fleet check

wooyongbin3-cpu:
## Summary

This PR fixes #97332 by only requiring fleet verification for gateway runtimes, not dashboard/serve runtimes.

## Problem

The updater was treating any runtime (dashboard, serve) as evidence that gateway fleet verification was needed. This caused `hermes update` to exit 1 even when only a dashboard was running and the update succeeded.

## Solution

Filter `pre_update_plan.runtimes` to only count gateway runtimes as evidence requiring fleet verification:

```python
has_gateway = any(
    getattr(r, 'kind', '') == 'gateway'
    for r in pre_update_plan.runtimes
)
if has_gateway:
    return True
```

Dashboard/serve-only plans no longer trigger the gateway fleet check.

## Changes

- Modified `hermes_cli/update_cmd.py`: Added gateway-specific filtering in `_fleet_probe_expected_runtimes`

---

🤖 Generated with Codebuff

### DIFF
--- a/hermes_cli/update_cmd.py
+++ b/hermes_cli/update_cmd.py
@@ -10103,7 +10103,14 @@ def _fleet_probe_expected_runtimes(
         return True
     try:
         if pre_update_plan is not None and pre_update_plan.runtimes:
-            return True
+            # Only gateway runtimes require fleet verification;
+            # dashboard/serve-only plans don't need gateway rows.
+            has_gateway = any(
+                getattr(r, 'kind', '') == 'gateway'
+                for r in pre_update_plan.runtimes
+            )
+            if has_gateway:
+                return True
     except Exception:
         pass
     return False


alt-glitch:

Duplicate of #97350: both make fleet verification expect rows only for gateway-kind plan records. #97350 is the earlier, test-covered implementation.

teknium1:
Same fix landed in #103478 (f58fcc8118) via the earlier #97350 by @liuhao1024 (cherry-picked with credit). Thanks for the report.

PR #86584, opened 2026-08-15
fix(kanban): tolerate string timestamps in _fmt_ts

magicbluesmoke:
## What

`hermes kanban show` crashes with `TypeError: 'str' object cannot be interpreted as an integer` when a task's `completed_at` (or `started_at`) is stored as a string.

A worker can write `completed_at` as an ISO-8601 datetime string (e.g. `'2026-08-04T21:05:15.679975'`) or an epoch encoded as a str. The DB column is INTEGER-typed, but upstream already anticipates string timestamps flowing through the codebase — `_to_epoch` in `kanban_db.py` explicitly accepts ISO-8601 and numeric strings. The display layer `_fmt_ts`, however, only accepts `int`, so a string reaching it crashes the `show` command.

## Change

Widen `_fmt_ts` to accept `str | int`, render ISO-8601 and epoch-as-str timestamps the same as ints, and add a regression test covering the crash input and related shapes (ISO string, space-separated string, epoch-as-str, int, `None`, `0`).

## Test

`pytest tests/hermes_cli/test_kanban_cli.py -q` — 5 passed, including the new `test_fmt_ts_tolerates_string_timestamps`.

PR #97462, opened 2026-08-28
fix(kanban): coerce foreign timestamp values when reading task/run/event rows

### LINKED ISSUE #97455: [Bug]: Non-epoch timestamps written into INTEGER columns make `kanban show` crash with `invalid literal for int()
## Summary

A write path in the Kanban feature persists **ISO-8601 strings** into SQLite columns that are declared and consumed as **unix-epoch integers**. Because SQLite does not enforce column affinity on insert, the write succeeds silently — but any subsequent `hermes kanban show <task_id>` for the affected card (JSON *and* plain text) crashes with:

```
kanban: invalid literal for int() with base 10: '2026-08-28T14:35:22+00:00'
```

`kanban list` keeps working board-wide; only `show` is broken, which makes the affected card effectively undiagnosable through the CLI. In my case the corruption appeared immediately after a worker executed the `kanban_request_changes` review transition, and I only recovered the card's state by querying `kanban.db` directly with SQL.

## Environment

- Hermes Agent v0.20.5 (2026.8.19), install method: git, local HEAD `f0c0c986`
- Windows 11 · Python 3.11.15 · SQLite 3.53.1 · WAL journal mode
- Multi-profile Kanban board (dispatcher + worker profiles, review lifecycle in active use)

## Affected columns (all observed in one board)

| Table | Column | Corrupted value (verbatim from DB) |
|---|---|---|
| `task_runs` | `ended_at` | `'2026-08-28T14:35:22+00:00'` |
| `task_events` | `created_at` | `'2026-08-28T16:58:56.613Z'` |
| `tasks` | `completed_at` | `'2026-08-28T16:58:56.613Z'` |

### Steps to Reproduce

Fastest deterministic path (no multi-agent setup needed), on a test board:

```sql
UPDATE task_runs SET ended_at = '2026-08-28T14:35:22+00:00' WHERE id = <some_run_id>;
```

then:

```
hermes kanban --board <board> show <task_id>
→ kanban: invalid literal for int() with base 10: '2026-08-28T14:35:22+00:00'
```

Natural path (how I hit it): run a card through implement → `kanban_request_review` → reviewer runs `kanban_request_changes` from a worker session → the transition writes an ISO string into `task_runs.ended_at` → every later `show` of that card crashes.

### Expected Behavior

- Every timestamp stored in Kanban DB columns (`task_runs.*`, `task_events.*`, `tasks.*`) is a unix-epoch integer, as declared in the schema.
- `hermes kanban show <task_id>` (and the Desktop card detail view) reads any card's run/event history without errors, regardless of what transitions produced it — including cards that went through worker-executed review transitions (`kanban_request_changes`).
- A malformed timestamp, if it ever reaches the DB, degrades gracefully (skipped/coerced with a warning), instead of crashing every subsequent read of the card.

### Actual Behavior

- A worker-executed `kanban_request_changes` transition writes ISO-8601 strings (`'2026-08-28T14:35:22+00:00'`, `'2026-08-28T16:58:56.613Z'`) into INTEGER timestamp columns (`task_runs.ended_at`, `task_events.created_at`, `tasks.completed_at`).
- Every `hermes kanban show <task_id>` for the affected card then crashes with `ValueError: invalid literal for int() with base 10` (raised at `kanban_db.py:1287 from_row`, via `list_runs`).
- The Desktop Kanban UI hits the same crash when rendering the card detail (`plugin_api.py:583 get_task`) — 42 occurrences in a 3-minute window while the UI retried.
- The card becomes undiagnosable through both the CLI and the UI; the only recovery is querying `kanban.db` directly with SQL and manually normalizing the timestamps.

### Affected Component

Other

N/A (CLI only)

### Debug Report

```shell
https://paste.rs/aQsHN
```

### Operating System

Windows 11

### Python Version

_No response_

### Hermes Version

_No response_

```shell

```

**1. Writer side (type violation).** The schema declares the columns as INTEGER (`hermes_cli/kanban_db.py:1471`):

```python
ended_at            INTEGER,
```

but at least one write path passes a `datetime.isoformat()` string instead of `int(time.time())`. SQLite stores it without complaint. In my board the corruption landed during the review lifecycle — three rows across two cards, all written on transitions (`changes_requested` / completion) executed by worker sessions. I did not bisect the exact call site, but the common denominator is transition-timestamp writes performed from worker-invoked tool paths.

**2. Reader side (fragile parse).** The run-row parser casts with a bare `int()` and no coercion (`hermes_cli/kanban_db.py:1287`):

```python
ended_at=(int(row["ended_at"]) if row["ended_at"] is not None else None),
```

`int('2026-08-28T14:35:22+00:00')` raises `ValueError: invalid literal for int() with base 10`, which propagates out of the CLI.

The frustrating part: **the codebase already contains the right tool** — `_to_epoch()` (`hermes_cli/kanban_db.py:11292`) handles ints, numeric strings, and ISO-8601 — but it is only used by `task_age()`. The run/event/task parsers never touch it.

### Proposed Fix (optional)

1. **Reader hardening (one line):** route all timestamp reads through `_to_epoch()` — `task_runs.started_at/ended_at/last_heartbeat_at`, `task_events.created_at`, `task_comments.created_at`, `tasks.created_at/started_at/completed_at/claim_expires`. This alone makes existing corrupted DBs readable again.
2. **Writer audit (root cause):** find the transition code path that produces `isoformat()` values and normalize it to epoch ints. Worth grepping the worker-invoked kanban tools specifically — that's where mine originated.
3. **Optional one-time migration:** normalize any existing non-integer timestamp values on DB open, or at minimum log a warning.

liuhao1024:
## What does this PR do?

`kanban show` (text and JSON), the `kanban_show` worker context, and the dashboard's run history all crash with `invalid literal for int() with base 10: '2026-08-28T14:35:22+00:00'` once a timestamp column in `tasks` / `task_runs` / `task_events` holds an ISO-8601 string, making the affected card unreadable (#97455).

SQLite's type affinity accepts those writes silently, and every mutator in `kanban_db` already writes `int(time.time())` — the stray strings come from rows written outside the mutators (e.g. a worker shell running raw SQL). Since the writer can't be pinned down, the fix is on the reader side: a new `_coerce_epoch` helper passes numeric values through, converts timezone-aware ISO-8601 strings (both `+00:00` and `Z` spellings observed in the wild) to epoch seconds, and degrades anything untrustworthy (naive ISO, garbage) to `None` so the card stays readable. Naive ISO strings are deliberately dropped rather than guessed at a timezone, so a malformed value is never rendered as a wrong time.

Applied at the row mappers (`Task.from_row`, `Run.from_row`, both `Event` constructors), which covers every consumer of those objects, plus the two raw-SQL reader spots in the dispatcher (respawn guard, rate-limit cooldown) and the worker-context "recent work" renderer that called `int()` directly.

## Related Issue

Fixes #97455

## Type of Change

## Changes Made

- `hermes_cli/kanban_db.py`: added `_coerce_epoch()` — epoch/float/epoch-as-str pass through, timezone-aware ISO-8601 converts to epoch seconds, naive ISO and garbage degrade to `None`
- `hermes_cli/kanban_db.py`: `Task.from_row` / `Run.from_row` / both `Event` constructions now coerce `created_at` / `started_at` / `completed_at` / `ended_at` / `claim_expires` / `last_heartbeat_at` instead of passing strays through or calling bare `int()`
- `hermes_cli/kanban_db.py`: `check_respawn_guard` (rate-limit cooldown + recent-completion window) and the `build_worker_context` "recent work" renderer coerce raw `ended_at` values instead of `int()`
- `tests/hermes_cli/test_kanban_db.py`: `_coerce_epoch` matrix test + row-mapper regression that seeds the exact corruption from the issue
- `tests/hermes_cli/test_kanban_cli.py`: end-to-end `kanban show` regression (text + JSON) reproducing the issue's `UPDATE` scenario

## How to Test

1. `uv run --extra dev python -m pytest tests/hermes_cli/test_kanban_db.py tests/hermes_cli/test_kanban_cli.py tests/tools/test_kanban_tools.py -q` — Observed result: `70 passed, 1 skipped` (the skip is a pre-existing environment-conditional test, unrelated)
2. `uv run --extra dev python -m pytest tests/hermes_cli/test_kanban_db.py::test_readers_coerce_iso_strings_in_integer_timestamp_columns tests/hermes_cli/test_kanban_cli.py::test_show_survives_iso_strings_in_integer_timestamp_columns -q` — Observed result: `2 passed`; the tests store `'2026-08-28T14:35:22+00:00'` into `task_runs.ended_at` and `'2026-08-28T16:58:56.613Z'` into `tasks.completed_at` / `task_events.created_at` exactly as in the issue, then assert `show` renders and the JSON payload carries epoch ints
3. On the issue's repro (`UPDATE task_runs SET ended_at = '2026-08-28T14:35:22+00:00' ...` then `hermes kanban show <task_id>`): should pass — the card renders and `runs[].ended_at` is the coerced epoch int `1756384522` instead of raising `invalid literal for int()`

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate (#86584 tolerates string timestamps only in the CLI `_fmt_ts` renderer — this crash happens earlier, at the `Run.from_row` data layer, which #86584 does not touch)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (behavior rationale documented in the `_coerce_epoch` docstring)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (pure-Python datetime parsing, no platform-specific code)

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -86,6 +86,7 @@ import logging
 import time
 from contextvars import ContextVar, Token
 from dataclasses import dataclass, field
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Iterable, Mapping, Optional
 
@@ -481,6 +482,45 @@ _CTX_MAX_COMMENTS       = 30      # most recent N comments shown in full
 _CTX_MAX_FIELD_BYTES    = 4 * 1024   # 4 KB per summary/error/metadata/result
 _CTX_MAX_BODY_BYTES     = 8 * 1024   # 8 KB per task.body (opening post)
 _CTX_MAX_COMMENT_BYTES  = 2 * 1024   # 2 KB per comment
+
+
+def _coerce_epoch(value: Any) -> Optional[int]:
+    """Coerce a timestamp-column value to a unix-epoch int, tolerating strays.
+
+    The timestamp columns of ``tasks`` / ``task_runs`` / ``task_events`` are
+    declared INTEGER, but SQLite's type affinity lets an ISO-8601 string (or
+    an epoch-as-str) land there silently when a row is written outside the
+    mutators — e.g. a worker shell running raw SQL (#97455). Every mutator in
+    this module writes ``int(time.time())``, so readers only need to tolerate
+    foreign rows: numeric values pass through, timezone-aware ISO-8601
+    strings are converted to epoch seconds, and anything else degrades to
+    ``None`` instead of crashing every later read of the card. Naive ISO
+    strings are NOT guessed at a timezone — an unparseable clock reading is
+    dropped rather than rendered as a wrong time.
+    """
+    if value is None:
+        return None
+    if isinstance(value, bool):
+        return None
+    if isinstance(value, (int, float)):
+        return int(value)
+    if isinstance(value, str):
+        s = value.strip()
+        if not s:
+            return None
+        try:
+            return int(s)
+        except ValueError:
+            pass
+        try:
+            iso = s[:-1] + "+00:00" if s.endswith("Z") else s
+            dt = datetime.fromisoformat(iso)
+        except ValueError:
+            return None
+        if dt.tzinfo is None:
+            return None
+        return int(dt.timestamp())
+    return None
 
 
 def _relative_age(ts: Optional[int], now: Optional[int] = None) -> str:
@@ -1162,15 +1202,15 @@ class Task:
             status=row["status"],
             priority=row["priority"],
             created_by=row["created_by"],
-            created_at=row["created_at"],
-            started_at=row["started_at"],
-            completed_at=row["completed_at"],
+            created_at=(_coerce_epoch(row["created_at"]) or 0),
+            started_at=_coerce_epoch(row["started_at"]),
+            completed_at=_coerce_epoch(row["completed_at"]),
             workspace_kind=row["workspace_kind"],
             workspace_path=row["workspace_path"],
             branch_name=row["branch_name"] if "branch_name" in keys else None,
             project_id=row["project_id"] if "project_id" in keys else None,
             claim_lock=row["claim_lock"],
-            claim_expires=row["claim_expires"],
+            claim_expires=_coerce_epoch(row["claim_expires"]),
             tenant=row["tenant"] if "tenant" in keys else None,
             result=row["result"] if "result" in keys else None,
             idempotency_key=row["idempotency_key"] if "idempotency_key" in keys else None,
@@ -1192,7 +1232,9 @@ class Task:
                 row["max_runtime_seconds"] if "max_runtime_seconds" in keys else None
             ),
             last_heartbeat_at=(
-                row["last_heartbeat_at"] if "last_heartbeat_at" in keys else None
+                _coerce_epoch(row["last_heartbeat_at"])
+                if "last_heartbeat_at" in keys
+                else None
             ),
             current_run_id=(
                 row["current_run_id"] if "current_run_id" in keys else None
@@ -1279,12 +1321,12 @@ class Run:
             step_key=row["step_key"],
             status=row["status"],
             claim_lock=row["claim_lock"],
-            claim_expires=row["claim_expires"],
+            claim_expires=_coerce_epoch(row["claim_expires"]),
             worker_pid=row["worker_pid"],
             max_runtime_seconds=row["max_runtime_seconds"],
-            last_heartbeat_at=row["last_heartbeat_at"],
-            started_at=int(row["started_at"]),
-            ended_at=(int(row["ended_at"]) if row["ended_at"] is not None else None),
+            last_heartbeat_at=_coerce_epoch(row["last_heartbeat_at"]),
+            started_at=(_coerce_epoch(row["started_at"]) or 0),
+            ended_at=_coerce_epoch(row["ended_at"]),
             outcome=row["outcome"],
             summary=row["summary"],
             metadata=meta,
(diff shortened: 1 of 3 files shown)


liuhao1024:
The Windows-only failure in this run is a known flaky timing test, unrelated to this diff.

**Failing test:** `tests/test_desktop_update_windows_progress.py::test_progress_advances_while_the_orchestrator_blocks` — `AssertionError: /progress unresponsive until deadline (last error: TimeoutError('timed out'))`.

**Why it's unrelated to this change:**
- This PR only touches `hermes_cli/kanban_db.py` and its two kanban test files — no shared code path with the Desktop update `/progress` socket/HTTP chain.
- This test has a documented flake history on CI: 6d1284a073 (Aug 19, *"deflake the Windows progress self-test (both race directions seen in CI)"*) and 18a15a46d8 (Aug 20, *"Windows progress self-test survives transient /progress socket stalls"*). The failure here (`/progress unresponsive until deadline`) is the same transient-stall shape the second deflake commit targeted, so a stall can still slip through under runner load.
- Everything else in the same run is green: the full `Python tests / Run tests` job passed, and 153/154 selected Windows tests passed — only this deadline-sensitive test timed out.

No rerun attempted since external contributors can't trigger one; happy to rebase if that's preferred.

KeyArgo:
Verified this covers the issue completely. Reproduced the crash independently: `UPDATE task_runs SET ended_at='2026-08-28T14:35:22+00:00'` then `hermes kanban show` raises `ValueError: invalid literal for int()` from `Run.from_row` (kanban_db.py:1287), exactly as reported. The reader-side `_coerce_epoch` approach is the right call — it matches the reporter's own proposed fix (route all timestamp reads through `_to_epoch()`), and this PR covers every affected column the issue enumerated (`task_runs.ended_at`, `task_events.created_at`, `tasks.completed_at`) plus the dispatcher/worker-context raw `int()` call sites. Nice to see both text and JSON `show` covered with the exact seeded corruption. LGTM.

liuhao1024:
Thanks for reproducing the crash independently and confirming the reader-side coercion covers the issue — matching the reporter's intent with a defensive read path was exactly the goal.

PR #97463, opened 2026-08-28
fix(desktop): apply app color scheme to inline previews

### LINKED ISSUE #95814: ::preview iframe defaults to light color-scheme, ignoring app theme
### Summary

Inline widgets rendered via `::preview{file="..."}` load in an iframe that does not inherit the host app's `color-scheme`. Under the dark theme, a widget renders as a bright white card against dark chat, unless the widget itself declares `:root { color-scheme: dark; }`.

### Steps to reproduce

1. Run Hermes desktop with the dark theme active.
2. Write a minimal HTML file with a transparent background and no explicit `color-scheme`.
3. Embed it in a chat message with `::preview{file="..."}`.
4. The widget renders with a white backdrop.

### Expected

The widget inherits the host app's active theme, and `prefers-color-scheme` resolves correctly inside the iframe.

### Actual

The iframe uses the UA default `color-scheme: light` regardless of host theme.

### Notes

- **A transparent background alone is not sufficient.** This was the obvious first fix and it does not work — the UA default still drives form controls, scrollbars, and the canvas beneath the transparent body.
- The working fix is `:root { color-scheme: dark; }` inside the widget, but that hardcodes dark and will be wrong for a user on the light theme.
- Guidance to widget authors is to use theme tokens (`var(--foreground)`, `var(--muted-foreground)`, `var(--accent)`, `var(--border)`, `var(--card)`), which implies the iframe is *meant* to be theme-aware. The `color-scheme` default looks like an oversight rather than a design choice.

### Suggested fix

Propagate the host's `color-scheme` to the preview iframe.

### Environment

| Field | Value |
|---|---|
| Hermes Agent | v0.20.5 (2026.8.19) |
| Upstream ref | `f0187332` |
| Install method | git |
| Python | 3.11.9 |
| OpenAI SDK | 2.24.0 |
| OS | macOS 26.5.2, dark theme |
| Model provider | GitHub Copilot |

BlackishGreen33:
## Related Issue

Fixes #95814

## What does this PR do?

Inline `::preview` HTML runs inside a sandboxed iframe. The iframe element already followed the Desktop theme, but the HTML document inside it did not declare a `color-scheme`. In dark mode, browser-drawn controls, scrollbars, and the transparent canvas could therefore stay light.

This change passes the app's resolved light or dark mode into the existing theme prelude that is injected into the preview document. The preview's own CSS still comes later, so a page that chooses its own color scheme keeps that choice.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- Add the resolved `light` or `dark` `color-scheme` to the inline preview document's existing `:root` theme prelude.
- Rebuild the preview document when the resolved color scheme changes, and reuse the same value on the iframe element.
- Add regression coverage for both light and dark preview documents.

## How to Test

The new regression failed on current `main` because the generated preview document contained no `color-scheme` declaration.

1. `npx vitest run --project ui src/components/assistant-ui/inline-preview-directive.test.ts` — 16 passed.
2. `npm run test:ui` — 679 test files passed, 6707 tests passed.
3. `npm run typecheck` — all three Desktop TypeScript checks passed.
4. `npm run lint` — completed with 0 errors (124 warnings outside the changed files).
5. `npx prettier --check src/components/assistant-ui/inline-preview-directive.tsx src/components/assistant-ui/inline-preview-directive.test.ts` and `git diff --check` — passed.

### Code

- [ ] I've run `pytest tests/ -q` and all tests pass — N/A for this Desktop-only TypeScript change; the full Desktop renderer suite is listed above
- [x] I've tested on my platform: macOS 26.5.1 arm64

- [x] I've updated relevant documentation — N/A, no user-facing workflow or setting changed
- [x] I've updated `cli-config.yaml.example` — N/A, no config keys changed
- [x] I've updated `CONTRIBUTING.md` or `AGENTS.md` — N/A, no architecture or workflow changed
- [x] I've considered cross-platform impact — this uses the existing browser CSS and React theme path without platform-specific APIs
- [x] I've updated tool descriptions/schemas — N/A, no model-facing tool or schema changed

## Screenshots / Logs

N/A — this changes the generated iframe document theme and does not alter layout. The light and dark outputs are covered by the regression test above.

### DIFF
--- a/apps/desktop/src/components/assistant-ui/inline-preview-directive.test.ts
+++ b/apps/desktop/src/components/assistant-ui/inline-preview-directive.test.ts
@@ -24,7 +24,7 @@ describe('directiveFrameHeight', () => {
 })
 
 describe('withInlineChrome', () => {
-  const prelude = themePrelude({ '--foreground': '#eee' }, 'Inter')
+  const prelude = themePrelude({ '--foreground': '#eee' }, 'Inter', 'dark')
 
   it('puts the theme prelude FIRST so page styles override it', () => {
     const doc = '<html><head><style>body{color:red}</style></head><body><h1>hi</h1></body></html>'
@@ -52,8 +52,16 @@ describe('withInlineChrome', () => {
 })
 
 describe('themePrelude', () => {
+  it.each(['light', 'dark'] as const)('sets the iframe document color scheme to %s', colorScheme => {
+    expect(themePrelude({}, '', colorScheme)).toContain(`color-scheme:${colorScheme}`)
+  })
+
   it('carries resolved tokens, transparent background, and the app font', () => {
-    const prelude = themePrelude({ '--foreground': 'oklch(0.9 0 0)', '--accent': '#7aa2f7' }, 'Inter, sans-serif')
+    const prelude = themePrelude(
+      { '--foreground': 'oklch(0.9 0 0)', '--accent': '#7aa2f7' },
+      'Inter, sans-serif',
+      'dark'
+    )
 
     expect(prelude).toContain('--foreground:oklch(0.9 0 0)')
     expect(prelude).toContain('--accent:#7aa2f7')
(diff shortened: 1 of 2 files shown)


OutThisLife:
Superseded by #122004 (or #122005 for the markdown fixes) — same approach landed as part of the p3-rated fix sweep, with the missing bug-class tests and full i18n coverage. Your work is credited there. Thanks!

PR #97466, opened 2026-08-28
fix(mcp): fully redact Bearer values in mcp test output

### LINKED ISSUE #97460: security: MCP test and dashboard errors can expose Bearer fragments
## Summary

Current `hermes mcp test` partially masks resolved Authorization values as `first4***last4`, exposing a reusable credential fragment. If an MCP probe exception includes an `Authorization: Bearer <value>` string, the CLI and dashboard test endpoint can return that exception text without credential redaction.

This complements #84106 (`config get`) and #24302 (URL query redaction); it concerns header display and probe-error rendering.

## Environment

- Hermes checkout v0.20.6, local HEAD `39f1e1881a027cead71c81b7b747be1c7ad6853a`
- current upstream `main` was also inspected and retains the relevant behavior

No real credential was used in this report.

## Synthetic reproduction findings

With a dummy `Authorization: Bearer SYNTHETIC_MCP_BEARER_NOT_A_SECRET_123456` header:

1. `cmd_mcp_test` resolves the header and prints first/last characters.
2. A mocked probe exception echoing that header is interpolated directly into `_error(...)`.
3. `POST /api/mcp/servers/{name}/test` returns `{"error": str(exc)}`.

The normal MCP runtime connect-error formatter already removes Bearer values fully, so behavior is inconsistent.

## Expected

- Authorization/header credentials must be replaced completely, not fingerprint-masked.
- Probe/config/dashboard error text must pass the same Bearer/header redactor before display or JSON return.
- Neither full value nor chosen prefix/suffix from the synthetic token should appear.
- `${ENV_VAR}` should remain the on-disk configuration representation; this fix should not require moving secrets into config.

## Suggested tests

Use one synthetic dummy and assert its full value, a unique prefix, and a unique suffix are absent from:

- successful `hermes mcp test` auth display
- CLI probe exceptions
- dashboard MCP test errors
- config diagnostics
- normal runtime connect errors/log capture

Also retain the raw `${ENV_VAR}` template and `0600` permissions for newly created secret files.

## Suggested implementation boundary

Add an MCP-specific full-redaction helper for credential header values and error strings. Do not reuse generic partial `mask_secret` for Authorization output. Apply generic Bearer-pattern redaction after structured header redaction as defense in depth.

686f6c61:
## What does this PR do?

`hermes mcp test` resolved Authorization headers and printed `first4***last4`, which is still a reusable credential fragment. Probe exceptions that echoed `Authorization: Bearer <value>` were interpolated into the CLI error line and into `POST /api/mcp/servers/{name}/test` as `"error": str(exc)`.

This fully replaces Authorization/Bearer credentials with `***` (no prefix/suffix) on those surfaces, then runs `redact_sensitive_text(..., force=True)` as a second pass. `${ENV_VAR}` templates stay on disk; the probe does not need secrets inlined in config.

## Related Issue

Fixes #97460

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] Security fix
- [ ] Documentation update

## Changes Made

- `hermes_cli/mcp_config.py`: `redact_mcp_probe_text`; stop fingerprint-masking MCP test auth display; redact CLI probe exceptions
- `hermes_cli/web_routers/mcp.py`: dashboard MCP test errors go through the same helper
- `tests/hermes_cli/test_mcp_probe_redaction.py`: synthetic Bearer absent from display, CLI errors, and dashboard JSON

## How to Test

```text
python -m pytest tests/hermes_cli/test_mcp_probe_redaction.py tests/hermes_cli/test_mcp_config.py::TestMcpTest -q
# 8 passed
```

Not runtime-tested against a real MCP server.

### Code

- [x] I've run the targeted pytest files above

## Screenshots / Logs

N/A

### DIFF
--- a/hermes_cli/mcp_config.py
+++ b/hermes_cli/mcp_config.py
@@ -32,6 +32,35 @@ logger = logging.getLogger(__name__)
 
 _ENV_VAR_NAME_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
 
+# MCP test/dashboard surfaces must not fingerprint Authorization values
+# (firstN/lastN is a reusable credential fragment). Replace the credential
+# entirely, then run the generic redactor as defense in depth.
+_AUTH_HEADER_VALUE_RE = re.compile(
+    r"((?:Proxy-)?Authorization:\s*(?:[A-Za-z][\w.+-]*\s+)?)([^\s\"']+)",
+    re.IGNORECASE,
+)
+_AUTH_SCHEME_VALUE_RE = re.compile(
+    r"\b((?:Bearer|Basic|Token|Digest)\s+)([^\s\"']+)",
+    re.IGNORECASE,
+)
+
+
+def redact_mcp_probe_text(text: object) -> str:
+    """Fully redact MCP probe/display strings before they leave the process.
+
+    Authorization/Bearer credentials become ``***`` (no prefix/suffix). The
+    generic secret redactor then runs with ``force=True`` so other credential
+    shapes in the same exception still cannot leak.
+    """
+    raw = "" if text is None else str(text)
+    if not raw:
+        return raw
+    redacted = _AUTH_HEADER_VALUE_RE.sub(lambda m: f"{m.group(1)}***", raw)
+    redacted = _AUTH_SCHEME_VALUE_RE.sub(lambda m: f"{m.group(1)}***", redacted)
+    from agent.redact import redact_sensitive_text
+
+    return redact_sensitive_text(redacted, force=True)
+
 
 _MCP_PRESETS: Dict[str, Dict[str, Any]] = {
     "codex": {
@@ -774,13 +803,12 @@ def cmd_mcp_test(args):
     elif headers:
         for k, v in headers.items():
             if isinstance(v, str) and ("key" in k.lower() or "auth" in k.lower()):
-                # Mask the value (accepts ${VAR} and Cursor-style ${env:VAR})
-                resolved = _ENV_VAR_PATTERN.sub(lambda m: os.getenv(_env_ref_name(m.group(1)), ""), v)
-                if len(resolved) > 8:
-                    masked = resolved[:4] + "***" + resolved[-4:]
+                # Keep ${ENV_VAR} templates as the display form. Never interpolate
+                # a resolved secret into the CLI — first4/last4 is still reusable.
+                if _ENV_VAR_PATTERN.search(v):
+                    print(f"    {k}: {v}")
                 else:
-                    masked = "***"
-                print(f"    {k}: {masked}")
+                    print(f"    {k}: {redact_mcp_probe_text(v)}")
     else:
         _info("Auth: none")
 
@@ -791,7 +819,7 @@ def cmd_mcp_test(args):
         elapsed_ms = (time.monotonic() - start) * 1000
     except Exception as exc:
         elapsed_ms = (time.monotonic() - start) * 1000
-        _error(f"Connection failed ({elapsed_ms:.0f}ms): {exc}")
+        _error(f"Connection failed ({elapsed_ms:.0f}ms): {redact_mcp_probe_text(exc)}")
         return
 
     _success(f"Connected ({elapsed_ms:.0f}ms)")
--- a/hermes_cli/mcp_config.py
+++ b/hermes_cli/mcp_config.py
@@ -43,6 +43,10 @@ _AUTH_SCHEME_VALUE_RE = re.compile(
     r"\b((?:Bearer|Basic|Token|Digest)\s+)([^\s\"']+)",
     re.IGNORECASE,
 )
+_AUTH_SCHEME_PREFIX_RE = re.compile(
+    r"^(?:Bearer|Basic|Token|Digest)\s+",
+    re.IGNORECASE,
+)
 
 
 def redact_mcp_probe_text(text: object) -> str:
@@ -60,6 +64,46 @@ def redact_mcp_probe_text(text: object) -> str:
     from agent.redact import redact_sensitive_text
 
     return redact_sensitive_text(redacted, force=True)
+
+
+def _header_value_is_only_env_refs(value: str) -> bool:
+    """True when every non-scheme span in *value* is a ``${VAR}`` reference."""
+    leftover = _ENV_VAR_PATTERN.sub("", value)
+    leftover = _AUTH_SCHEME_PREFIX_RE.sub("", leftover)
+    return leftover.strip(" \t;,") == ""
+
+
+def redact_mcp_header_display(name: str, value: object) -> str:
+    """Fail-closed CLI display for a credential-shaped MCP header.
+
+    A value that is only ``${ENV}`` references (optionally with an auth scheme
+    word) may be shown — it carries no secret. Anything else, including opaque
+    API keys and mixed template+literal strings, is replaced with ``***``.
+    ``name`` stays paired with the value so callers cannot drop the header
+    identity before this policy runs.
+    """
+    raw = "" if value is None else str(value)
+    if raw and _header_value_is_only_env_refs(raw):
+        return raw
+    # Pair name+value for the generic redactor, then still fail closed — an
+    # opaque token with no recognized scheme must not print.
+    redact_mcp_probe_text(f"{name}: {raw}")
+    return "***"
+
+
+def _redact_probe_exception(exc: BaseException) -> Exception:
+    """Return a raise-able exception whose ``str()`` is safe to print."""
+    root = _unwrap_exception_group(exc)
+    safe = redact_mcp_probe_text(root)
+    if safe == str(root) and isinstance(root, Exception):
+        return root
+    try:
+        rebuilt = type(root)(safe)
+        if str(rebuilt) == safe and isinstance(rebuilt, Exception):
+            return rebuilt
+    except Exception:
+        pass
+    return RuntimeError(safe)
 
 
 _MCP_PRESETS: Dict[str, Dict[str, Any]] = {
@@ -423,7 +467,7 @@ def _probe_single_server(
     try:
         _run_on_mcp_loop(_probe(), timeout=connect_timeout + 10)
     except BaseException as exc:
-        raise _unwrap_exception_group(exc) from None
+        raise _redact_probe_exception(exc) from None
     finally:
         _stop_mcp_loop_if_idle()
 
@@ -594,7 +638,7 @@ def cmd_mcp_add(args):
     try:
         tools = _probe_single_server(name, server_config)
     except Exception as exc:
-        _error(f"Failed to connect: {exc}")
+        _error(f"Failed to connect: {redact_mcp_probe_text(exc)}")
         if _confirm("Save config anyway (you can test later)?", default=False):
             server_config["enabled"] = False
             if _save_mcp_server(name, server_config):
@@ -803,12 +847,9 @@ def cmd_mcp_test(args):
     elif headers:
         for k, v in headers.items():
             if isinstance(v, str) and ("key" in k.lower() or "auth" in k.lower()):
-                # Keep ${ENV_VAR} templates as the display form. Never interpolate
-                # a resolved secret into the CLI — first4/last4 is still reusable.
-                if _ENV_VAR_PATTERN.search(v):
-                    print(f"    {k}: {v}")
-                else:
-                    print(f"    {k}: {redact_mcp_probe_text(v)}")
+                # Keep header identity with the value. A ${ENV} substring is
+                # not proof the rest of the header is non-secret.
+                print(f"    {k}: {redact_mcp_header_display(k, v)}")
     else:
         _info("Auth: none")
 
@@ -931,7 +972,7 @@ def _reauth_oauth_server(name: str, server_config: dict) -> bool:
             )
         except Exception:
             humanized = None
-        _error(f"Authentication failed: {humanized or exc}")
+        _error(f"Authentication failed: {redact_mcp_probe_text(humanized or exc)}")
         return False
 
 
@@ -1037,7 +1078,7 @@ def cmd_mcp_configure(args):
     try:
         all_tools = _probe_single_server(name, cfg)
     except Exception as exc:
-        _error(f"Failed to connect: {exc}")
+        _error(f"Failed to connect: {redact_mcp_probe_text(exc)}")
         return
 
     if not all_tools:
--- a/hermes_cli/mcp_config.py
+++ b/hermes_cli/mcp_config.py
@@ -47,20 +47,33 @@ _AUTH_SCHEME_PREFIX_RE = re.compile(
     r"^(?:Bearer|Basic|Token|Digest)\s+",
     re.IGNORECASE,
 )
+# Opaque API-key headers — fully replace the value (no head/tail fingerprint).
+_MCP_SECRET_HEADER_RE = re.compile(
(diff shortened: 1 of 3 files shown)


andrexibiza (commented):
Reviewing exact head `24e704ca6608ecc8c337c84356df3f32bc397001` against `main@e60983a69730c058ce772829df3273aee6de3889` (the commit is a direct child of that main object). I read the actual three-file diff and the full behavior-critical MCP config/dashboard paths, checked the issue/competing implementation history, and verified the hosted acceptance for this exact commit: [CI 33213770196](https://github.com/NousResearch/hermes-agent/actions/runs/33213770196), [Docker 33213769480](https://github.com/NousResearch/hermes-agent/actions/runs/33213769480), and [Nix 33213769492](https://github.com/NousResearch/hermes-agent/actions/runs/33213769492) all succeeded. Attribution is also clean: `github@00b.tech` resolves to `686f6c61` on current main.

The direction is strong. Replacing reusable Authorization fragments with a non-reusable sentinel is the correct boundary, the structured Authorization/scheme pass is materially stronger than relying on the generic log redactor alone, and covering both classic CLI output and the dashboard JSON response is exactly the right pair of public surfaces.

I would not merge this exact object yet, though, because two credential-boundary holes remain.

**P1 — the new redactor is applied at selected consumers instead of the shared probe failure boundary, so sibling CLI commands still emit the raw exception.**

At this head `cmd_mcp_test()` now does:

```py
_error(f"Connection failed (...): {redact_mcp_probe_text(exc)}")
```

but the same `_probe_single_server()` failure is still interpolated unchanged in the other probe consumers in `hermes_cli/mcp_config.py`:

```py
# cmd_mcp_add
except Exception as exc:
    _error(f"Failed to connect: {exc}")

# cmd_mcp_login re-probe
except Exception as exc:
    _error(f"Failed to connect: {exc}")
```

So an HTTP/MCP exception containing an Authorization or API-key header is protected when the operator runs `hermes mcp test`, but can still be disclosed by `hermes mcp add` or the login re-probe. That is the same secret crossing the same probe boundary; command selection must not decide whether it is safe to print.

This is also the unique useful semantic from the now-closed competing implementation [#97470](https://github.com/NousResearch/hermes-agent/pull/97470): liuhao1024 explicitly closed that PR in favor of this one, and in the [closure note](https://github.com/NousResearch/hermes-agent/pull/97470#issuecomment-5458124809) identified these two sibling consumers and proposed redacting once at `_probe_single_server`'s existing `_unwrap_exception_group` seam. Please absorb that contribution here rather than leaving the fix split across consumer-specific catches. The regression needs at least `mcp add` and `mcp login` witnesses with a synthetic credential-bearing exception; ideally the invariant is simpler: every exception leaving the probe helper is already safe for display.

**P1 — `cmd_mcp_test` discards the header identity before redaction, so opaque API-key headers can still print, and any `${ENV}` occurrence makes the entire value trusted.**

The branch first correctly decides that a header is credential-shaped from `k`:

```py
if isinstance(v, str) and ("key" in k.lower() or "auth" in k.lower()):
```

but for a literal value it then calls `redact_mcp_probe_text(v)` — only the value, not `k`. The helper's full-replacement rules recognize `Authorization:` or an auth scheme (`Bearer|Basic|Token|Digest`). The generic `agent.redact` API-key-header rule also needs the header name to be present. Therefore a perfectly valid configuration such as:

```yaml
headers:
  X-Api-Key: opaquecredential1234567890
```

can reach the display helper as the bare string `opaquecredential1234567890`, with no header context left to trigger full replacement. This is particularly important because #97460 is about credential **headers**, not only Bearer Authorization.

The template branch has the inverse proof problem:

```py
if _ENV_VAR_PATTERN.search(v):
    print(f"    {k}: {v}")
```

Finding one environment reference proves only that one span is a reference; it does not prove every other literal span in the same header is non-secret. A mixed value such as `Bearer ${MCP_TOKEN}; backup=<literal-secret>` is therefore printed verbatim whenever the placeholder survives config loading.

Please keep the key/value pair together through the display policy and fail closed for every credential-shaped header: either render only verified-safe env-reference names or render the value as the full redaction sentinel. Add an opaque `X-Api-Key` witness and a mixed `${VAR}` + literal witness so this cannot regress back into partial trust.

Lineage/ownership is otherwise clean and should stay explicit: [#97460](https://github.com/NousResearch/hermes-agent/issues/97460) is JuliaF1988's originating security report; #97470 is now a superseded/closed parallel attempt whose shared-probe observation should be credited and folded into this carrier, not landed separately. I do not see a reason for two redaction authorities here — one MCP probe/display policy, applied before any credential-bearing diagnostic crosses a public output boundary, is the safer shape.

The hard part is already here: you replaced fingerprint masking with actual redaction and gave it focused tests. Close these two remaining escape routes and this becomes a much stronger security fix. 🚀

686f6c61:
Fixed in 1f15fef70.

Probe exceptions are redacted once at the `_probe_single_server` raise seam (the sibling-consumer gap @liuhao1024 called out on #97470), so `hermes mcp add`, `mcp login`, and `mcp configure` print already-safe text. `cmd_mcp_test` still keeps header identity with the value and fail-closes: only a pure `${ENV}` template (optional scheme word) is shown; opaque `X-Api-Key` values and mixed template+literal strings become `***`.

```text
python -m pytest tests/hermes_cli/test_mcp_probe_redaction.py
# 13 passed
```

andrexibiza (commented):
Re-reviewed exact head `1f15fef708bbe28313d7093a6032983368eda161`.

The follow-up closes the two original Authorization-path findings: `_probe_single_server` now owns the exception-redaction boundary, and successful CLI display keeps the header identity while failing closed for literal or mixed-template values. Exact-head [CI](https://github.com/NousResearch/hermes-agent/actions/runs/33218320870), [Docker](https://github.com/NousResearch/hermes-agent/actions/runs/33218320158), and [Nix](https://github.com/NousResearch/hermes-agent/actions/runs/33218320364) are green.

### P1 — the shared error path still fingerprints opaque API-key headers

[`redact_mcp_probe_text`](https://github.com/686f6c61/hermes-agent/blob/1f15fef708bbe28313d7093a6032983368eda161/hermes_cli/mcp_config.py#L35-L66) fully replaces only `[Proxy-]Authorization` and scheme-token shapes. Every `X-Api-Key` / `Api-Key` / token-header family falls into the generic pass. That pass recognizes those headers, but routes their values through [`_mask_token`](https://github.com/686f6c61/hermes-agent/blob/1f15fef708bbe28313d7093a6032983368eda161/agent/redact.py#L590-L609), which deliberately preserves six leading and four trailing characters ([matcher](https://github.com/686f6c61/hermes-agent/blob/1f15fef708bbe28313d7093a6032983368eda161/agent/redact.py#L342-L353), [application](https://github.com/686f6c61/hermes-agent/blob/1f15fef708bbe28313d7093a6032983368eda161/agent/redact.py#L930-L934)).

Deterministic witness at this head:

```python
redact_mcp_probe_text(
    "connect failed: X-Api-Key: opaquecredential1234567890ABCDEF"
)
# => "connect failed: X-Api-Key: opaque...CDEF"
```

So the value leaving the shared seam is still reusable credential material: `mcp test`, `add`, `login`, `configure`, and dashboard JSON can expose the same prefix/suffix fragment class #97460 requires removed. The matcher is also token-shaped rather than header-value-shaped: `Authorization: Digest username="u", response="opaquecredential1234567890ABCDEF"` leaves the `response` credential intact after masking only `username=`.

The [new suite](https://github.com/686f6c61/hermes-agent/blob/1f15fef708bbe28313d7093a6032983368eda161/tests/hermes_cli/test_mcp_probe_redaction.py#L9-L100) exercises `X-Api-Key` only through the success-display helper; every exception witness still uses `Authorization: Bearer ...`. Its shared assertion also has prefix/suffix sentinels only for `SYNTHETIC`, not `OPAQUE_API_KEY`, so an opaque-key exception test using that helper would still pass on `opaque...CDEF`.

Close the class at the MCP egress policy: fully replace the complete value of every credential header Hermes recognizes before the generic redactor runs, then add a direct `redact_mcp_probe_text` opaque-key witness, one shared-seam or dashboard/CLI opaque-key exception witness, and distinct opaque-key prefix/suffix assertions (or an exact `X-Api-Key: ***` assertion).

This remains a security blocker.

686f6c61:
Fixed in ecbce7fd7.

`redact_mcp_probe_text` now fully replaces opaque API-key header values (`X-Api-Key`, `Api-Key`, …) and Digest `username=` / `response=` parameters with `***` before the generic redactor runs, so prefix/suffix fingerprints cannot leave the shared probe seam. Added a direct helper witness, a `_probe_single_server` exception witness, and prefix/suffix assertions on the opaque key.

```text
python -m pytest tests/hermes_cli/test_mcp_probe_redaction.py
# 16 passed
```

andrexibiza (commented):
Re-reviewed exact head `ecbce7fd7bcec1d5af2e3679b1d42a53a3d97426` against the PR's recorded base `e60983a69730c058ce772829df3273aee6de3889`.

The follow-up closes the prior single-token wire-format witness: `X-Api-Key: opaquecredential...` is now fully replaced before the generic masking pass, the new assertions reject the opaque prefix and suffix, and exact-head [CI](https://github.com/NousResearch/hermes-agent/actions/runs/33224705227), [Docker](https://github.com/NousResearch/hermes-agent/actions/runs/33224704832), and [Nix](https://github.com/NousResearch/hermes-agent/actions/runs/33224704913) all succeeded.

One security blocker remains.

### P1 — the egress policy is still token-, order-, and serialization-dependent, so complete credential values can leave the shared seam

In [`redact_mcp_probe_text`](https://github.com/NousResearch/hermes-agent/blob/ecbce7fd7bcec1d5af2e3679b1d42a53a3d97426/hermes_cli/mcp_config.py#L38-L79), `_AUTH_HEADER_VALUE_RE` runs before `_AUTHZ_PARAM_RE` and consumes only the first unquoted token. For a structured Authorization value, that first pass can delete the parameter key while leaving its quoted value orphaned outside every later matcher.

Deterministic witness against the exact substitution order at this head:

```python
redact_mcp_probe_text(
    'Authorization: Digest response="opaquecredential1234567890ABCDEF", username="u"'
)
# => 'Authorization: Digest ***"opaquecredential1234567890ABCDEF", username=***'
```

The full `response` credential survives. The new [Digest test](https://github.com/NousResearch/hermes-agent/blob/ecbce7fd7bcec1d5af2e3679b1d42a53a3d97426/tests/hermes_cli/test_mcp_probe_redaction.py#L83-L90) puts the benign `username="u"` parameter first and the synthetic `response` second, which avoids the failing ordering. It also passes while the first parameter's value remains in the output as `***"u"`; the assertion proves only that the second parameter was removed, not that Digest parameters are fully redacted.

The opaque-header rule has the same shape problem: it recognizes only a bare wire-format `X-Api-Key:` followed by one `\S+` token. A common serialized exception representation bypasses both the MCP rule and the generic header rule:

```python
redact_mcp_probe_text(
    "headers={'X-Api-Key': 'opaquecredential1234567890ABCDEF'}"
)
# => "headers={'X-Api-Key': 'opaquecredential1234567890ABCDEF'}"
```

The helper is the shared exception egress boundary; no caller can prove that an external exception will serialize headers in exactly one bare-line format or that a structured auth parameter will appear only after a sacrificial first parameter.

Close the class at field identity rather than token shape: once a recognized credential header key is found, replace its complete associated value independent of scheme, quoted/unquoted form, parameter order, and wire/JSON/Python mapping serialization. Add a permutation matrix with each sensitive Digest parameter first/middle/last, bare and quoted header keys/values, and direct-helper plus `_probe_single_server`/dashboard witnesses. The credential-header vocabulary should also come from one authority (or the generic redactor should expose a full-replacement mode), rather than copying a second list that can drift back to partial masking.

This remains a security blocker.

686f6c61:
Fixed in 6173d9ab8.

`redact_mcp_probe_text` now replaces the complete associated value of a recognized credential header (`Authorization` / `Proxy-Authorization` and `agent.redact._SECRET_HEADER_NAMES`) independent of scheme, quoting, Digest parameter order, and wire/JSON/Python mapping serialization. The Digest-response-first and `headers={'X-Api-Key': '…'}` witnesses no longer retain the opaque value.

Added a permutation matrix (each sensitive Digest parameter first/middle/last, quoted and unquoted values, mixed mapping quotes) plus `_probe_single_server` and dashboard witnesses.

`pytest tests/hermes_cli/test_mcp_probe_redaction.py` — 40 passed.

andrexibiza (commented):
Re-reviewed exact head `6173d9ab8b9cbc45fc82558963d1dd2b9a7181fa` against current `main@a2e19d484cb5591df8dafe667c93345b62d9bf06` and the full four-commit submitted train.

The credential-boundary blockers from my prior reviews are closed on this object.

- Probe failures are now sanitized once at `_probe_single_server` before they escape to any caller; `mcp add`, `mcp login`, `mcp configure`, `mcp test`, and the dashboard all consume an already-safe exception, with consumer-side redaction retained as defense in depth.
- Successful header display keeps field identity through policy. Opaque credential-shaped values fail closed to `***`; a `${ENV}` occurrence no longer blesses adjacent literal material; only values composed entirely of env references plus an optional auth scheme are displayed.
- `redact_mcp_probe_text` now derives the opaque-header vocabulary from `agent.redact._SECRET_HEADER_NAMES` and fully replaces the associated value before the generic redactor runs. The regression matrix covers Authorization, opaque `X-Api-Key`, Digest with sensitive parameters in multiple orders/quoting forms, JSON/Python mapping serialization, the shared probe seam, dashboard JSON, and sibling CLI consumers.

I also checked the release object rather than transferring the original head's receipt. Every surviving commit has the required hosted matrix green:

- `24e704ca6608ecc8c337c84356df3f32bc397001` — CI `33213770196` ✅, Docker `33213769480` ✅, Nix `33213769492` ✅
- `1f15fef708bbe28313d7093a6032983368eda161` — CI `33218320870` ✅, Docker `33218320158` ✅, Nix `33218320364` ✅
- `ecbce7fd7bcec1d5af2e3679b1d42a53a3d97426` — CI `33224705227` ✅, Docker `33224704832` ✅, Nix `33224704913` ✅
- `6173d9ab8b9cbc45fc82558963d1dd2b9a7181fa` — CI `33230326648` ✅, Docker `33230326164` ✅, Nix `33230326143` ✅

Current `main` is 59 commits beyond the PR's recorded base; the full base→main compare is path-disjoint from all three submitted files (`hermes_cli/mcp_config.py`, `hermes_cli/web_routers/mcp.py`, `tests/hermes_cli/test_mcp_probe_redaction.py`), and GitHub currently reports the PR mergeable. There are no unresolved inline review threads.

One lineage point should stay visible in the landing history: closed #97470 by @liuhao1024 independently identified the shared `_probe_single_server` raise seam and the sibling `mcp add` / `mcp login` consumers; this carrier has now absorbed that semantic. #97460 remains JuliaF1988's originating report.

I do not see a remaining source, exact-object CI, or landing-drift blocker on this head.

teknium1:
Salvaged onto current main in #108695 (your four commits' final diff re-applied post-decomposition; authorship preserved via Co-authored-by). The base here predated the mcp_config/web_routers split, so a mechanical rebase wasn't viable — the redaction helpers, raise-seam placement, and fail-closed header display carried over intact, with the test monkeypatch seams repointed to tools.mcp_tool_loop / tools.mcp_tool_lifecycle. Thanks @686f6c61!

686f6c61:
#108695 is the current-main salvage of this redaction (helpers + raise-seam + fail-closed header display). Happy to close this if maintainers take that one.

PR #97467, opened 2026-08-28
fix(memory): omit current_entries on batch abort

### LINKED ISSUE #97316: memory: batch abort echoes the entire store back, so consolidation retries grow the context they were meant to shrink
### Summary

`memory` batch operations are all-or-nothing, and the abort path returns `current_entries` — the full contents of the store — alongside the error. For a large consolidation batch this makes each failed attempt cost tens of KB of context, and since the usual reason to run such a batch is *that the store is too full*, the retry loop inflates the very context it was invoked to reduce.

### What we measured

An agent tried to consolidate `MEMORY.md` (reported at ~98% of its char budget) with a single batch of 47+ operations. One `remove` had stale `old_text`:

```json
{
  "success": false,
  "error": "Operation 47 (remove): no entry matched '<snip>'. No operations were applied (batch is all-or-nothing).",
  "current_entries": [ ... ],
  "usage": "<n>/<limit>"
}
```

Payload breakdown:

| field | size |
|---|---|
| `error` | 115 chars |
| `current_entries` | **17,806 chars** |
| everything else | ~20 chars |
| **total tool response** | **18,015 chars** |

The tool call itself was 18,693 chars, so one failed consolidation round-trip cost ~37KB of context. Observed effect on that session: context reached 111K tokens, a single request took 94.7s, and the turn eventually died on a stale-stream timeout. From the user's side the agent simply stopped answering.

### Why the echo is right for single ops and wrong for batches

For a single failed `replace`/`remove`, echoing entries is clearly correct and the error message says so: *"Check current_entries below and retry with the exact text of the entry you want to replace."* The agent needs the exact strings.

For a batch, the same payload is mostly waste:

- **All-or-nothing means nothing changed**, so the agent already has an accurate picture from the previous response — nothing was applied that it needs re-syncing on.
- **Only one operation failed.** The 46 that validated fine don't need their entries re-sent.
- **The batch is the consolidation path**, i.e. the one invoked precisely when the store is large, which is when the echo is largest.

### Suggested fix

Any of these would break the loop; the first is the smallest change:

1. **On batch abort, return only what failed** — the operation index, its `old_text`, and the nearest existing entries (say the top 3 by similarity) instead of the whole store. Keeps the retry actionable at a fraction of the size.
2. **Validate the whole batch and report every failure at once.** Today it aborts at the first bad operation, so an agent fixing N stale `old_text` values needs N round-trips, each paying the full echo. One pass listing all mismatches converges in one retry.
3. **Cap or gate the echo** — omit `current_entries` above some size, or add an opt-out flag for callers that already hold the state.

(2) is independently valuable: first-failure-only reporting is what makes the retry count scale with the number of stale references.

### Environment

- Hermes Agent v0.20.6 (`2026.8.27`)
- `memory.memory_char_limit` raised well above the default, which is what allows a store big enough for an 18KB echo — but the same dynamic applies at the default cap, just proportionally smaller.
- Reproduces with any batch where one `old_text` is stale; the store size sets the cost.

686f6c61:
## What does this PR do?

`memory` batch ops are all-or-nothing, but the abort path returned `current_entries` — the full store — next to the error. A consolidation batch is used precisely when the store is large, so each stale `old_text` retry paid tens of KB back into the turn.

Single `replace`/`remove` misses still echo entries (the error text tells the model to use them). Batch aborts now return the failed operation + `usage` only.

## Related Issue

Fixes #97316

## Type of Change

- [ ] Security fix
- [ ] Documentation update

## Changes Made

- `tools/memory_tool.py`: `_batch_error` and the batch budget-overflow path omit `current_entries`
- `tests/tools/test_memory_tool.py`: batch abort does not echo store contents; single replace still does

## How to Test

```text
python -m pytest tests/tools/test_memory_tool.py -q
# 43 passed
```

### Code

- [x] I've run the targeted pytest file above

## Screenshots / Logs

N/A

### DIFF
--- a/tests/tools/test_memory_tool.py
+++ b/tests/tools/test_memory_tool.py
@@ -216,11 +216,38 @@ class TestMemoryConsolidationGracefulDegrade:
         for _ in range(cap):
             r = store.apply_batch("memory", bad_batch)
             assert r["success"] is False
-            assert "current_entries" in r  # still actionable under cap
+            assert "current_entries" not in r
+            assert "No operations were applied" in r["error"]
         r = store.apply_batch("memory", bad_batch)
         assert r["success"] is False
         assert r["done"] is True
         assert "continue with your reply" in r["error"]
+        assert "current_entries" not in r
+
+    def test_apply_batch_abort_does_not_echo_store(self, store):
+        """A failed consolidation must not pay the whole MEMORY.md back (#97316)."""
+        store.add("memory", "fact A that is unique and long enough to matter")
+        store.add("memory", "fact B stays in the store after the abort")
+        result = store.apply_batch(
+            "memory",
+            [{"action": "remove", "old_text": "this substring is not in any entry"}],
+        )
+        assert result["success"] is False
+        assert "current_entries" not in result
+        payload = json.dumps(result)
+        assert "fact A that is unique" not in payload
+        assert "fact B stays in the store" not in payload
+        assert "No operations were applied" in result["error"]
+        assert store.memory_entries == [
+            "fact A that is unique and long enough to matter",
+            "fact B stays in the store after the abort",
+        ]
+
+    def test_single_replace_still_echoes_entries_on_miss(self, store):
+        store.add("memory", "fact A")
+        result = store.replace("memory", "nonexistent", "new")
+        assert result["success"] is False
+        assert result["current_entries"] == ["fact A"]
 
     def test_success_and_turn_boundary_reset_failure_budget(self, store):
         store.add("memory", "real entry")
--- a/tools/memory_tool.py
+++ b/tools/memory_tool.py
@@ -594,7 +594,9 @@ class MemoryStore:
 
         Semantics: all-or-nothing. If any op is malformed, doesn't match, or
         the net result would exceed the char limit, NOTHING is written and an
-        error is returned describing the first failure plus the live state.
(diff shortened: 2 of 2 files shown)


kshitijk4poor:
Triage note (perf/P2 label sweep): the change trims the `current_entries` echo from memory batch-abort responses — a response-payload shape fix, ~40 lines, applies cleanly to current main. It is perf-adjacent (context/token cost on the LLM tool path) but P2 overstates it; suggesting P3 and leaving it to the memory-tool reviewers. No judgment on merits here.

Benchmark (macOS; `MemoryStore` filled to ~90% of its char limit with 200-char entries, then a batch `add` that would exceed the limit; `len(json.dumps(result))` ×5, median): main `30b83ab7b1` **2,081 bytes** → PR head **205 bytes**, `current_entries` absent. What actually improved: a failed consolidation retry no longer re-sends the whole store into the context it was invoked to shrink (~10× smaller error payload on this fixture; scales with store size).

Script: `bench/97467.py <checkout>` (fills store, triggers abort, measures payload).

kshitijk4poor:
Thanks @686f6c61 — this landed on main via #117106 (merge 6882320d59), cherry-picked from this branch so your commit authorship is preserved, rebased onto current main with the review follow-ups on top. #97316 is fixed by that merge. Closing this one as superseded; the credit is yours.

PR #97469, opened 2026-08-28
fix(desktop): keep primary session selector in sync on model switch

### LINKED ISSUE #97425: [Bug]: Desktop live model switch leaves primary session selector stale
## Summary

In an existing idle Desktop conversation, a successful model/provider switch can update the global composer selection while the visible primary session selector remains on the previous model/provider.

This is a focused follow-up to #68876. It isolates the renderer state-sync defect from the confirmation, runtime-rebind, transcript, and backend deferred-turn work.

## Steps to reproduce

1. Open Hermes Desktop and start a conversation with provider A / model X.
2. Send a message and wait until the session is idle.
3. Use the model picker in the same conversation to select provider B / model Y.
4. Wait for the session-scoped `config.set` request to acknowledge successfully.
5. Observe the model picker, composer pill, and active session selector.

## Actual behavior

The global composer selection can contain provider B / model Y while the active primary session view and visible composer pill still contain provider A / model X until a later session metadata refresh.

The current code has two relevant paths:

- `apps/desktop/src/app/chat/session-view.tsx` makes the active `$sessionStates[runtimeId]` slice authoritative for `PRIMARY_SESSION_VIEW.$model` and `$provider` once a runtime exists.
- `apps/desktop/src/app/session/hooks/use-model-controls.ts` updates `$currentModel` / `$currentProvider` for a primary selection, but does not synchronously update the active runtime's session slice.

`session.info` intentionally does not blindly write the global composer atoms because stale heartbeats and profile defaults must not overwrite manual selections. That guard should remain intact.

## Expected behavior

After an acknowledged successful switch on an idle existing primary session:

- the active session view and composer pill show the acknowledged provider/model;
- the model picker selection and visible pill agree;
- a failed or declined switch restores the prior session and composer state;
- a later stale `session.info` heartbeat does not overwrite a newer manual composer selection.

## Scope

This issue is intentionally limited to the Desktop renderer path for an **idle existing primary session** and the provider/model identity displayed by that session.

Likely files are the model-control/session-state write path plus focused renderer tests. The fix should preserve the existing separation between draft/global composer state and live per-runtime session state.

## Non-goals

Please do not bundle these into this issue:

- `confirm_required` confirmation UX, already handled by #92492/#95635;
- runtime re-keying/rebinding after a model-switch rebuild, handled by #93942/#95600;
- deferred busy-turn scheduling or rate-limit retry behavior;
- transcript hydration or message ordering;
- typed `/model <name>` reconciliation, tracked by #80904/#81112;
- profile-default persistence or stale new-chat provider state, tracked separately by #62055/#96063;
- gateway model-switch or provider credential changes.

## Regression coverage

Add a component/state regression that seeds an active runtime session slice with the old provider/model, performs an acknowledged picker selection, and asserts the rendered `PRIMARY_SESSION_VIEW`/`ModelPill` changes to the new provider/model. Also cover failure/confirmation rollback and stale heartbeat protection without changing tile isolation behavior.

## Related work

- #68876 — broad provider/model switch synchronization report
- #90796 — closed/superseded acknowledgement repaint redesign
- #92492 / #95635 — confirmation handshake
- #93942 / #95600 — rebuilt-runtime stream rebind
- #80904 / #81112 — typed `/model` synchronization

686f6c61:
## What does this PR do?

On an idle existing Desktop conversation, a successful model/provider switch updated `$currentModel` / `$currentProvider` while `PRIMARY_SESSION_VIEW` kept reading `$sessionStates[runtimeId]`. The visible selector and composer pill could stay on the previous pair until a later session metadata refresh.

`selectModel` now paints the live runtime slice on success (via the session-tile delegate, or `$sessionStates` when the delegate is not registered). Failure and `confirm_required` still roll that slice back. Tile picks still do not rewrite the primary composer atoms. `session.info` still does not write the global composer atoms.

## Related Issue

Fixes #97425

## Type of Change

- [ ] Security fix
- [ ] Documentation update

## Changes Made

- `apps/desktop/src/app/session/hooks/use-model-controls.ts`: paint/rollback the live session slice on primary (and tile) switches
- `apps/desktop/src/app/session/hooks/use-model-controls.test.tsx`: acknowledged switch, failure rollback, confirm handshake, stale composer atoms, tile isolation

## How to Test

```text
cd apps/desktop
npx vitest run src/app/session/hooks/use-model-controls.test.tsx
# 26 passed
```

Not runtime-tested in the packaged Desktop app.

### Code

- [x] I've run the targeted vitest file above

## Screenshots / Logs

N/A

### DIFF
--- a/apps/desktop/src/app/session/hooks/use-model-controls.test.tsx
+++ b/apps/desktop/src/app/session/hooks/use-model-controls.test.tsx
@@ -21,8 +21,8 @@ import {
   $sessionStates,
   publishSessionState,
   sessionTileDelegate,
-  setSessionTileDelegate,
-  type SessionTileDelegate
+  type SessionTileDelegate,
+  setSessionTileDelegate
 } from '@/store/session-states'
 
 import { deferred } from '../../../test/deferred'
@@ -92,6 +92,7 @@ function installSessionDelegate() {
     updateSession: (runtimeId, updater) => {
       const previous =
         sessionStateCache.get(runtimeId) ?? $sessionStates.get()[runtimeId] ?? createClientSessionState()
+
       const next = updater(previous)
 
       if (next === previous) {
@@ -693,6 +694,7 @@ describe('useModelControls', () => {
     const requestGateway = vi.fn(async () => {
       throw new Error('no such model')
     })
+
     let controls!: Controls
 
     render(<Harness onReady={value => (controls = value)} requestGateway={requestGateway} />)
@@ -714,6 +716,7 @@ describe('useModelControls', () => {
     const requestGateway = vi.fn(async () => {
       throw new Error('no such model')
     })
+
     let controls!: Controls
 
     render(<Harness onReady={value => (controls = value)} requestGateway={requestGateway} />)
@@ -847,6 +850,7 @@ describe('useModelControls', () => {
     setCurrentModel('primary/model')
     setCurrentProvider('openai')
     seedRuntimeSlice('primary-runtime', 'primary/model', 'openai')
+
     seedRuntimeSlice('tile-runtime', 'old-tile', 'nous')
 
     const requestGateway = vi.fn(async () => ({ key: 'model', value: 'tile-model' }) as never)
--- a/apps/desktop/src/app/session/hooks/use-model-controls.test.tsx
+++ b/apps/desktop/src/app/session/hooks/use-model-controls.test.tsx
@@ -2,7 +2,9 @@ import { QueryClient } from '@tanstack/react-query'
 import { act, cleanup, render, renderHook, waitFor } from '@testing-library/react'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
+import { PRIMARY_SESSION_VIEW } from '@/app/chat/session-view'
 import { getGlobalModelInfo } from '@/hermes'
+import { createClientSessionState } from '@/lib/chat-runtime'
 import { modelOptionsQueryKey } from '@/lib/model-options'
 import { $activeGatewayProfile } from '@/store/profile'
 import {
@@ -14,6 +16,7 @@ import {
   setCurrentModelSource,
   setCurrentProvider
 } from '@/store/session'
+import { $sessionStates, publishSessionState } from '@/store/session-states'
 import type * as SessionStates from '@/store/session-states'
 
 import { deferred } from '../../../test/deferred'
@@ -78,10 +81,19 @@ function Harness({
   return null
 }
 
+function seedRuntimeSlice(runtimeId: string, model: string, provider: string) {
+  publishSessionState(runtimeId, {
+    ...createClientSessionState(`stored-${runtimeId}`),
+    model,
+    provider
+  })
+}
+
 describe('useModelControls', () => {
   beforeEach(() => {
     $activeGatewayProfile.set('default')
     $activeSessionId.set(null)
+    $sessionStates.set({})
     setCurrentModel('')
     setCurrentModelSource('')
     setCurrentProvider('')
@@ -92,6 +104,7 @@ describe('useModelControls', () => {
     vi.restoreAllMocks()
     $activeGatewayProfile.set('default')
     $activeSessionId.set(null)
+    $sessionStates.set({})
     setCurrentModel('')
     setCurrentModelSource('')
     setCurrentProvider('')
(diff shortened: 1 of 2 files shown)


thelonewander3r:
Thanks for picking this up — the diagnosis matches what I filed, and painting the
runtime slice rather than only the composer atoms is the right shape for the fix.
Four things I'd want resolved before this lands, one behavioral and three about
what the new tests actually prove.

### 1. The primary rollback restores from the wrong source

`prevModel` / `prevProvider` are captured from the composer atoms on the
`touchesPrimary` branch:

```ts
const prevModel = touchesPrimary ? $currentModel.get() : ($sessionStates.get()[liveSessionId!]?.model ?? '')
```

`rollbackSelection` now feeds those into `paintLiveSlice`, so on a failure or a
`confirm_required` the slice is repainted with whatever the *composer* held, not
with what the slice held. When the two disagree — which is the exact condition
this issue exists to describe — a failed switch moves the visible selector to a
model the session isn't running. Before this PR the primary rollback never
touched the slice, so failure was a no-op there.

The tile branch already gets this right by reading `$sessionStates[liveSessionId]`.
I'd capture the slice's own prior pair once, up front, and roll back to that:

```ts
const prevSlice = liveSessionId ? $sessionStates.get()[liveSessionId] : undefined
// ...
paintLiveSlice(prevSlice?.model ?? prevModel, prevSlice?.provider ?? prevProvider)
```

That keeps a declined or failed switch a true no-op on the surface the user is
looking at, which is what the "failed or declined switch restores the prior
state" bullet was asking for.

### 2. All five new tests exercise a branch production never takes

`use-model-controls.test.tsx` mocks the delegate away:

```ts
vi.mock('@/store/session-states', async importOriginal => ({
  ...(await importOriginal<typeof SessionStates>()),
  sessionTileDelegate: () => null
}))
```

so every new assertion lands on the new `$sessionStates` fallback inside
`paintLiveSlice`. In the app the delegate is always registered
(`use-session-tile-delegate.ts`, `updateSession: (runtimeId, updater) => updateSessionState(runtimeId, updater)`),
so the shipped path is `delegate.updateSession` → `updateSessionState`
(`use-session-state-cache.ts`), which additionally runs `persistInFlightTurnState`,
`publishSessionState`, and `syncSessionStateToView` → `syncRuntimeMetadataToView`.
That last one rewrites `$currentModel`, `$currentProvider`,
`$currentReasoningEffort`, `$currentServiceTier`, `$currentFastMode`,
`$yoloActive` and `$currentPersonality` from the slice for the active session.

None of that is covered by the 26 green tests. Could a couple of these register a
real delegate (or a `updateSession` that routes through `updateSessionState`) so
the acknowledged-switch and rollback cases run the path that actually ships?

### 3. The fallback isn't equivalent to the delegate path

A direct `publishSessionState` write bypasses `sessionStateCache`, and
`updateSessionState` takes its previous state from that cache via
`ensureSessionState`. So if the fallback ever runs for real — delegate not yet
registered during early mount — the cache still holds the old model and the next
`session.info` heartbeat republishes it, undoing the paint.

If the fallback exists only to keep the tests green with the delegate mocked out,
I'd rather drop it and fix the mock (see point 2 above). If it's meant as a real safety net,
it should go through the cache too, otherwise the two branches diverge in a way
the tests can't see.

### 4. The stale-heartbeat test doesn't exercise a heartbeat

`does not let a later composer-atom-only heartbeat hide a newer primary pick`
calls `setCurrentModel` / `setCurrentProvider` directly and then asserts
`PRIMARY_SESSION_VIEW`. That's asserting `primaryField`'s precedence rule, not
heartbeat protection. A real `session.info` writes the *slice* through
`updateSessionState` (`sessionInfoStatePatch` carries model/provider) and syncs
back into the composer atoms via `syncRuntimeMetadataToView` — so the "a later
stale `session.info` heartbeat does not overwrite a newer manual composer
selection" bullet is still uncovered. Worth either driving a real patch through
`updateSessionState` in that test or renaming it to what it checks.

None of this is a disagreement with the approach — 1 is a small capture change and
2–4 are about the tests describing the shipped path. Happy to look again once
they're addressed.

686f6c61:
Fixed in b314dfd4b.

Rollback now captures the live slice pair up front and restores that, not the composer atoms — a failed/declined switch is a no-op on the selector the user is looking at even when composer and slice disagreed. Writes go only through `sessionTileDelegate().updateSession` (no `$sessionStates` fallback). The new tests register a delegate that routes `updateSession` through a cache + `publishSessionState` + the focused-session metadata sync, matching the shipped `updateSessionState` path. The old composer-atom test is renamed to what it actually checks; a session.info-shaped patch is applied through the same `updateSession` path.

```text
cd apps/desktop
npx vitest run src/app/session/hooks/use-model-controls.test.tsx
# 28 passed
```

686f6c61:
Fixed in 586e9cc37.

CI `JS & TS checks` failed because `isStale` treated a missing session slice as a changed model, so the model-menu confirm handshake never sent the `confirm_expensive_model` resend. Confirm now compares composer atoms when there is no live slice. Also sorted the test named imports for eslint.

```text
cd apps/desktop
npx vitest run src/app/session/hooks/use-model-controls.test.tsx src/app/shell/model-menu-panel.test.tsx
# 50 passed
```

Issue #69377, opened 2026-07-22
multiplex_profiles: cron jobs registered in a secondary profile's store never fire — the single gateway only runs the default profile's ticker


marknisk:
Deployment: v2026.7.7.2 (0.18.2), Docker, s6-supervised gateway,
`multiplex_profiles: true`, secondary profile `home-ops` owning a Telegram
adapter.

The multiplexer serves the secondary profile's messaging lanes (config,
SOUL, model, memory all scope correctly), but the cron subsystem does not
follow: the gateway process starts ONE `InProcessCronScheduler` bound to the
process-global `HERMES_HOME`, so only `/opt/data/cron/jobs.json` is ticked.
A job registered from a secondary-profile session lands in
`<profile>/cron/jobs.json`, reports a valid `next_run_at` — and never runs.
Our secondary profile's `cron/ticker_heartbeat` had not advanced in 15 days
while the default profile's beat every minute.

The agent-facing surface actively misleads here: the profile-scoped cronjob
tool accepts the registration, validates the script path against the
profile's `scripts/` dir, and shows the job as `scheduled` with a concrete
`next_run_at`. Nothing warns that no ticker owns that store.

Starting a second gateway for the profile is (correctly) refused by
`_guard_named_profile_under_multiplexer` — it would double-bind the
messaging platforms. So under multiplexing there is no supported way to get
a secondary profile's cron jobs to fire.

Suggestion: when `multiplex_profiles` is on, the default gateway should
start a scheduler (or tick pass) per served profile — mirroring how config
resolution already works per adapter-owning profile. Interim alternatives:
have the cronjob tool refuse/warn on registration into a store no ticker
owns, and surface per-profile ticker liveness in `hermes doctor`.

Workaround we run: duplicate the job as a no-agent script job in the
default profile's store, with a wrapper under the default `scripts/` dir
exec'ing the (self-anchored) profile script.

smfworks:
After investigating the code, this requires starting a cron ticker per served profile under multiplexing — a substantial architectural change to the scheduler lifecycle. The gateway currently starts one `InProcessCronScheduler` bound to the process-global `HERMES_HOME`; adding per-profile tickers needs careful design around:

1. Per-profile `HERMES_HOME` override during the tick pass
2. Scheduler lifecycle management (start/stop per profile)
3. Delivery routing (which adapter sends the result)
4. Thread safety with the multiplexer's existing profile scoping

[...]

PR #97484, opened 2026-08-28
fix(cron): profile_routes rescue for satellite-profile delivery preflight

### LINKED ISSUE #97476: Cron preflight false-blocks Telegram delivery for satellite profiles using gateway.profile_routes
## Summary

`cron/scheduler.py::_preflight_check_delivery` blocks a cron job's Telegram delivery even when the job would actually succeed at send time, because the preflight check consults the *wrong profile's* gateway config when the job lives inside a satellite (non-default) profile that uses `gateway.profile_routes` for delivery.

## Setup

- Multi-profile deployment with `gateway.multiplex_profiles: true` and `gateway.profile_routes` configured on the main gateway, e.g.:

```yaml
gateway:
  multiplex_profiles: true
  profile_routes:
    - name: grant-topic
      platform: telegram
      chat_id: '-1004306455751'
      thread_id: '14'
      profile: grant
```

- Satellite profile `grant` intentionally has **no** `TELEGRAM_BOT_TOKEN` of its own (correct — it shares the main bot via `profile_routes`, and giving it its own token previously caused a `duplicate_credential` fatal error against the primary profile's gateway).
- A cron job created *inside* the `grant` profile (`hermes --profile grant cron add ...` or equivalent) sets `deliver: "telegram:-1004306455751:14"`.

## Bug

On every scheduled run, the job is blocked before any inference call with:

```
⛔ Cron 'grant-weekly-synthesis' blocked by configuration validation (no LLM call was made): delivery platform 'telegram' has no gateway credentials configured (not connected). Configure it via hermes setup or change the job's deliver target. This alert is sent once; the job stays blocked until the configuration is fixed.
```

`last_status` is recorded as `blocked_config`.

## Root cause

`_preflight_check_delivery` (cron/scheduler.py, ~line 4826) calls:

```python
from gateway.config import load_gateway_config
gateway_config = load_gateway_config()
connected = {p.value for p in gateway_config.get_connected_platforms()}
```

This loads gateway config scoped to the profile the cron scheduler is currently running under (`grant`, in this case). Since `grant` has no Telegram credentials of its own by design, `get_connected_platforms()` correctly reports Telegram as not connected *for that profile* — but this ignores `gateway.profile_routes` entirely, which is the actual mechanism the main gateway uses to route `telegram:-1004306455751:14` deliveries to the `grant` profile's content without `grant` needing its own bot token.

The check never consults `profile_routes` (parsed via `gateway/profile_routing.py::parse_profile_routes`) before declaring the platform unreachable, so any job living inside a `profile_routes`-delivered satellite profile is permanently false-blocked, even though the exact same `deliver` string would succeed if the main gateway actually attempted the send.

## Impact

Any cron job created inside a satellite profile that delivers via `profile_routes` (rather than a Telegram token owned by that profile) is unconditionally blocked, every run, with a misleading error suggesting the platform simply isn't connected — when in fact routing is correctly configured, just not consulted by this specific check.

## Suggested fix

`_preflight_check_delivery` should check `gateway.profile_routes` (matching platform + chat_id/thread_id + the job's own profile) as an additional path to "connected," alongside the existing `get_connected_platforms()` check — the same way delivery resolution itself does at send time. Alternatively, load the *main* gateway's config (not the satellite profile's) when evaluating connectivity for a `profile_routes`-eligible platform.

## Environment

- Hermes v0.20.5
- Multi-profile Bot Mode deployment, 5 profiles (default + 4 satellites) sharing one Telegram bot via `profile_routes`
- `gateway.multiplex_profiles: true`

liuhao1024:
## What does this PR do?

Under `gateway.multiplex_profiles`, a satellite profile's cron jobs are ticked by the **primary** gateway's in-process ticker (#69377) and delivered through the **primary** gateway's live adapters — the satellite home intentionally holds no platform credentials of its own (installing a second token is a `duplicate_credential` fatal). `_preflight_check_delivery` loads the gateway config of the job's **own** home, so a platform delivered via `profile_routes` correctly reads as unconnected there, and the job is permanently blocked before any LLM call with a misleading "no gateway credentials configured (not connected)" error (#97476) — even though the exact same `deliver` string would be delivered fine by the primary's adapters.

This adds a `profile_routes` rescue: when the own-home config reports a platform unconnected, the preflight consults the **primary** home's `profile_routes`; an enabled route matching the platform that points at the profile currently being served means the delivery is the primary gateway's to make, so the check passes. The primary `config.yaml` is read directly (both the top-level and nested `gateway.` forms, mirroring `load_gateway_config`'s precedence) rather than via `load_gateway_config()`, so no primary platform config is loaded into the satellite process's environment. Lookup errors, missing configs, and disabled/foreign routes fail closed — the existing block stands.

## Related Issue

Fixes #97476

## Type of Change

## Changes Made

- `cron/scheduler.py` — new `_delivery_platform_routed_from_primary_gateway()` helper: resolves the primary home via `get_default_hermes_root()`, short-circuits when the current home IS the primary, parses `profile_routes` (top-level + nested form) via the existing `gateway.profile_routing.parse_profile_routes`, and matches enabled routes on platform + `profile_matches_home` (the current scheduler home, i.e. the exact mechanism the multiplex ticker scopes ticks with).
- `cron/scheduler.py` — `_preflight_check_delivery` now calls the helper in the not-connected branch before returning the block, mirroring the existing relay-fronted rescue's placement.
- `tests/cron/test_cron_multiplex_profile_route_preflight.py` — 6 regression tests: routed satellite passes, unrouted platform still blocked, route for another profile does not rescue, primary home itself skips the lookup, missing primary config fails closed, disabled route is inert.

## How to Test

1. `uv run --extra dev python -m pytest tests/cron/test_cron_multiplex_profile_route_preflight.py -q` → **Observed result: 6 passed.**
2. `uv run --extra dev python -m pytest tests/cron/test_cron_relay_delivery_guards.py tests/cron/test_cron_bot_chat_delivery.py -q` → **Observed result: 27 passed** (the existing relay/bot-chat preflight behavior is unchanged).
3. `uv run --extra dev python -m pytest tests/cron/test_scheduler.py -q` → **Observed result: 103 passed.**
4. Manual shape of the rescued path (covered by test 1): primary `config.yaml` with `gateway.profile_routes: [{platform: telegram, profile: grant, ...}]`, home override set to the `grant` profile, own-home gateway config reporting nothing connected → `_preflight_check_delivery({"deliver": "telegram:-100...:14"})` returns `None` (previously returned the "not connected" block).

### Code

- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — or N/A (no new keys; `profile_routes` already exists)

### DIFF
--- a/cron/scheduler.py
+++ b/cron/scheduler.py
@@ -5146,6 +5146,66 @@ def _preflight_check_provider_key(job: dict, cfg: dict) -> Optional[str]:
     return None
 
 
+def _delivery_platform_routed_from_primary_gateway(platform_name: str) -> bool:
+    """True when the primary gateway routes this platform to the profile the
+    scheduler is currently serving.
+
+    Under ``gateway.multiplex_profiles`` a satellite profile's cron jobs are
+    ticked by the primary gateway's in-process ticker (#69377) and delivered
+    through the primary gateway's live adapters — the satellite home never
+    holds the platform credentials itself (giving it a token of its own is a
+    ``duplicate_credential`` fatal). ``_preflight_check_delivery`` loads the
+    gateway config of the job's OWN home, where such a platform correctly
+    reads as unconnected; consulting the primary home's ``profile_routes``
+    keeps routed satellite jobs from being permanently false-blocked (#97476).
+    Reads the primary config.yaml directly (both the top-level and nested
+    ``gateway.`` forms) instead of ``load_gateway_config()`` so no primary
+    platform config leaks into this process's environment.
+    """
+    try:
+        from hermes_constants import get_default_hermes_root, get_hermes_home
+
+        primary_home = get_default_hermes_root()
+        current_home = Path(get_hermes_home())
+        if (
+            primary_home.expanduser().resolve(strict=False)
+            == current_home.expanduser().resolve(strict=False)
+        ):
+            return False  # this IS the primary home — nothing to consult
+        config_path = primary_home.expanduser() / "config.yaml"
+        if not config_path.exists():
+            return False
+
+        import yaml
+
+        with open(config_path, encoding="utf-8") as f:
+            raw = yaml.safe_load(f) or {}
+        routes_raw = raw.get("profile_routes")
+        if routes_raw is None and isinstance(raw.get("gateway"), dict):
+            routes_raw = raw["gateway"].get("profile_routes")
+        if not isinstance(routes_raw, list):
+            return False
+
+        from gateway.profile_routing import parse_profile_routes
+        from hermes_cli.profiles import profile_matches_home
+
+        platform_key = platform_name.lower()
+        for route in parse_profile_routes(routes_raw):
+            if (
+                route.enabled
+                and str(route.platform).lower() == platform_key
+                and profile_matches_home(route.profile)
+            ):
+                return True
+        return False
+    except Exception:
+        logger.debug(
+            "preflight: primary-gateway profile-route lookup unavailable",
+            exc_info=True,
+        )
+        return False
+
+
 def _preflight_check_delivery(job: dict) -> Optional[str]:
     """Check the job's delivery target(s) resolve to configured platforms.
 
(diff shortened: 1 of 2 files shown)


teknium1:
Merged via #99375 (merged, commit 9eb832aad7) — cherry-picked with your authorship preserved, including the test file. Earliest correct fix for the preflight false-block, and the targeted profile_routes read (fail-closed, no config leak into the satellite env) was the right call over loading the primary gateway config wholesale. Resolves #97476. Thanks!

PR #94062, opened 2026-08-24
fix(models): make the live Ollama Cloud listing authoritative over models.dev

liuhao1024:
## What does this PR do?

`fetch_ollama_cloud_models()` unioned models.dev-only IDs onto the live `/v1/models` listing. models.dev lags Ollama's retirements, so retired models (the reporter's examples: `kimi-k2.5`, `deepseek-v4-pro`, `deepseek-v4-flash`, `minimax-m2.5` — 4 phantom entries vs the 19-model live listing) persisted in the `/model` picker indefinitely. Selecting one fails at runtime with `Model '<name>' was not found in this provider's model listing`, and manually deleting the IDs from `ollama_cloud_models_cache.json` only helps until TTL expiry, when the next refresh re-unioned models.dev.

This makes the live listing authoritative for *existence*: when the live probe returns a non-empty list, models.dev no longer adds IDs of its own (the merge dedupe/suffix-stripping is unchanged). models.dev keeps its intended role as the full fallback when the live probe yields no models at all (no key, probe failure, empty listing) — that path is unchanged.

Also per the issue: the docstring's hardcoded "< 1 hour" now references `_OLLAMA_CLOUD_CACHE_TTL`, and the resolution-order notes match the actual merge semantics.

## Related Issue

Fixes #94041

## Type of Change

- [x] 🐛 Bug fix

## Changes Made

- `hermes_cli/models.py` (`fetch_ollama_cloud_models`):
  - In the merge loop, a models.dev entry that normalizes to an ID not already in the live list is skipped whenever `live_models` is non-empty, with a comment naming the retired-model failure mode (#94041). When `live_models` is empty the previous fallback behaviour (full models.dev list) is preserved.
  - Docstring updated: cache TTL now referenced via the constant; steps 3/4 describe models.dev as a full fallback rather than a gap-filler that adds IDs.
- `tests/hermes_cli/test_ollama_cloud_provider.py`:
  - `test_merges_live_and_models_dev` rewritten as `test_live_listing_authoritative_models_dev_only_entries_dropped` — same fixture, now pins that models.dev-only IDs stay out while live order/dedupe is preserved (docstring notes it supersedes the old union pin).
  - New `test_retired_ids_not_readded_on_refresh_after_cache_clear` — the reporter's TTL loop: a cache pre-seeded with a phantom ID plus a live listing without it must end with both the returned list and the persisted cache free of the phantom ID.

## How to Test

[...]

PR #94055, opened 2026-08-24
fix(ollama-cloud): hide retired models from picker

fangliquanflq:
## What does this PR do?

Ollama Cloud's model picker can list retired models that the live API no longer serves, causing users to select entries that fail provider validation. This change makes a successful `/v1/models` response authoritative, using models.dev only when live discovery is unavailable, and caches authoritative empty catalogs so stale fallback entries cannot reappear.

### Symptom

When models.dev still indexes a retired Ollama Cloud model, the picker includes it even though the live API omits it. Selecting that entry fails with `Model '<name>' was not found in this provider's model listing`.

### Impact

Users with Ollama Cloud credentials can be offered unusable models after cache refreshes until models.dev removes the retired entries.

### Bug Cause

**Trigger:** `hermes_cli/models.py:6268` / `fetch_ollama_cloud_models()` / a successful live response alongside models.dev-only entries

**Causal chain:**
1. The picker refreshes the Ollama Cloud catalog from `/v1/models`.
2. `fetch_ollama_cloud_models()` appends every models.dev-only entry to the successful live result.
3. Retired IDs are saved to disk and remain selectable even though provider validation rejects them.

**Why it is wrong:** models.dev is a secondary registry and can lag provider retirements, so it cannot define model existence when the provider's live catalog is available.

**Working sibling / contrast:** When live discovery is unavailable, models.dev remains a useful fallback and continues to populate the picker.

**Ruled out:** Suffix normalization was not the cause. `_strip_ollama_cloud_suffix()` prevents duplicate `:cloud` and `-cloud` aliases but intentionally leaves unrelated models.dev-only IDs unchanged.

### Fix

Return the de-duplicated live catalog immediately after a successful probe and persist it as the authoritative cache value. Only consult models.dev when the live probe is unavailable. Treat a successful empty response as authoritative and cacheable so an old non-empty fallback cannot be resurrected.

## Related Issue

Closes #94041

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `hermes_cli/models.py` - make successful Ollama Cloud discovery authoritative and reserve models.dev for fallback.
- `tests/hermes_cli/test_ollama_cloud_provider.py` - cover retired models.dev entries and authoritative empty live catalogs.

## How to Test

1. Run the Ollama Cloud provider regression suite.
2. Confirm the authoritative-live test returns only IDs from the mocked live API and never calls models.dev.
3. Confirm an empty live catalog remains empty after a cache read.

```bash
[...]

Issue #94041, opened 2026-08-24
ollama-cloud: /model picker lists retired models because fetch_ollama_cloud_models() unions live API + models.dev without filtering


mueblesindustria-design:
## Summary

The ollama-cloud model picker lists models that no longer exist in the live Ollama Cloud API. Selecting one of them fails at runtime with `Model '<name>' was not found in this provider's model listing`. Root cause: `fetch_ollama_cloud_models()` treats models.dev as a *secondary gap-filler* and unions it onto the live API list without ever intersecting against the live list — so any model models.dev still indexes (but Ollama has retired) persists in the picker indefinitely.

## Environment

- hermes-agent (git-installed, current main as of 2026-08-24)
- Linux arm64 (Debian 13, Raspberry Pi 5)
- `model.provider: ollama-cloud`, `model.base_url: https://ollama.com/v1`

## Evidence

Live API (`GET https://ollama.com/v1/models`, valid `OLLAMA_API_KEY`): **19 models**.
Disk cache `$HERMES_HOME/ollama_cloud_models_cache.json`: **23 models**.

Phantom entries present in cache/picker but absent from the live API:

- `kimi-k2.5` (replaced by `kimi-k2.6`)
- `deepseek-v4-pro` (untagged; replaced by `deepseek-v4-pro:0813`)
- `deepseek-v4-flash` (untagged; replaced by `deepseek-v4-flash:0731`)
- `minimax-m2.5` (replaced by `minimax-m2.7` / `minimax-m3`)

Manually deleting the four IDs from the cache file only helps until TTL expiry: on the next refresh the union with models.dev re-adds them.

## Root cause

`hermes_cli/models.py::fetch_ollama_cloud_models()` (docstring at line ~6232):

> Resolution order:
> 1. Disk cache (if fresh, < 1 hour, and not force_refresh)
> 2. Live /v1/models endpoint (primary — freshest source)
> 3. models.dev registry (secondary — fills gaps for unlisted models)
> 4. Merge: live models first, then models.dev additions (deduped)

Step 4 appends every models.dev entry that isn't already in the live list (`_strip_ollama_cloud_suffix` only normalizes `:cloud`/`-cloud` suffixes — it does not drop retired IDs). models.dev lags Ollama's catalog when providers retire models, so retired IDs survive in the merged result forever. The live API should be authoritative over *existence*; models.dev should only enrich metadata for models the live API already lists.

## Reproduction

[...]

Issue #81952, opened 2026-08-08
Corrupt profile config silently falls back to a paid OpenRouter default model and auto-ingests OPENROUTER_API_KEY — real-money spend without operator consent


cruzanstx:
## Summary

When a profile `config.yaml` fails to parse, non-interactive one-shot runs (`hermes --profile X chat -Q`) do not fail. They continue with no config, silently select the built-in default `z-ai/glm-5.2` on `openrouter` (`PREFERRED_SILENT_DEFAULT_MODEL`, `hermes_cli/models.py:1381`), auto-ingest `OPENROUTER_API_KEY` from the profile `.env` into the credential pool, and run the full agentic job (100+ turns) at real cost. The corrupt-config handler (`_backup_corrupt_config`, `hermes_cli/config.py:45`) snapshots the file to `config.yaml.corrupt.<ts>.bak` but **leaves the broken file in place**, so every subsequent scheduled run repeats the spend. The parse failure itself is never written to `errors.log` — the only breadcrumbs are an INFO line and, when the auxiliary lane engages, a WARNING.

In our deployment, a single mis-indented line (introduced by a scripted config edit) put 4 profiles in this state; cron-driven one-shots then burned ~$10 on OpenRouter across 3 days before we noticed — via the OpenRouter billing dashboard, not via Hermes. Session accounting recorded the spend correctly, but no surface alerted the operator.

## Environment

- hermes_cli 0.20.0 (venv install), Linux, Python 3.x
- Non-interactive one-shot mode: `hermes --profile <name> chat -Q -q "..." --max-turns 120` from cron
- `OPENROUTER_API_KEY` present in profile `.env` (kept for image-gen plugins)
- Profile config names `model.provider: openai-codex`; OpenRouter never configured as a chat provider

## Reproduction

1. In a profile `config.yaml` that sets e.g. `model.provider: openai-codex`, break the YAML anywhere (e.g. under-indent one mapping line).
2. Ensure `OPENROUTER_API_KEY` is in the profile `.env` and not present in the auth store's `suppressed_sources`.
3. Run `hermes --profile <name> chat -Q -q "hello"`.
4. Observe:
   - `config.yaml.corrupt.<ts>.bak` appears; the broken `config.yaml` stays active.
   - agent.log: `INFO cli: No model configured — defaulting to z-ai/glm-5.2 for provider openrouter` (`hermes_cli/cli_agent_setup_mixin.py:169`)
[...]

PR #97487, opened 2026-08-28
fix(provider): consult current provider's live catalog before OpenRouter fallback in detect_provider_for_model

sj-unit72:
## Problem

On the ACP path, a model whose bare name exists in OpenRouter's catalog can silently hijack to OpenRouter even when the **current provider** already serves it.

`acp_adapter/server.py#_resolve_model_selection` runs `parse_model_input` → `('ollama-cloud', X)`, and because provider == current it calls `detect_provider_for_model`. That function consulted **only the static `_PROVIDER_MODELS` catalog** then fell through to `_find_openrouter_slug`. `ollama-cloud` is not in the static list, so any ollama-cloud model whose bare name also exists in OpenRouter (e.g. `glm-5.3-flash` → `zai/glm-5.3-flash`) returned `('openrouter', zai/...)`. `set_session_model` treats that as `provider_changed=True` and **rebuilds the agent on OpenRouter**, silently billing OpenRouter credits for a model the current provider already serves.

Confirmed live: three sessions ran `z-ai/glm-5.3-flash` on `openrouter.ai/api/v1`, ~$0.29 estimated / $0.875 billed, while the ollama-cloud catalog *does* serve the model — it was just never consulted.

## Fix

Add a step between the static-catalog check and the OpenRouter fallback that consults the **current provider's live catalog** via `provider_model_ids()` (normalized with `_strip_ollama_cloud_suffix`). If the current provider actually serves the model, return `None` (stay put) instead of falling through to OpenRouter. Wrapped in `try/except` so a live-fetch failure degrades to the old fallback.

## Verification

- `detect_provider_for_model("glm-5.3-flash", "ollama-cloud")` → `None` before fix: `('openrouter', 'zai/glm-5.3-flash')` (the hijack)
- after fix → `None` (stays on ollama-cloud) ✓
- Genuine remaps still resolve: `gpt-4o` → `openai`, `deepseek-chat` → `deepseek` ✓

This fixes the whole class: any currently-served model can no longer be re-routed to OpenRouter merely because its bare name appears there.

### DIFF
--- a/hermes_cli/models.py
+++ b/hermes_cli/models.py
@@ -3538,6 +3538,21 @@ def detect_provider_for_model(
     if _model_in_provider_catalog(name.lower(), _provider_keys(current_provider)):
         return None
 
+    # --- Step 1.5: the CURRENT provider's live catalog is authoritative ---
+    # A model the current provider actually serves (e.g. ollama-cloud serving
+    # glm-5.3-flash, which is not in the static _PROVIDER_MODELS list) must NOT
+    # be re-routed to OpenRouter merely because its bare name also appears in
+    # the OpenRouter catalog. This fixed the silent hijack where an
+    # ollama-cloud session was rebuilt on OpenRouter (billing via
+    # openrouter.ai) because the live catalog was never consulted.
+    try:
+        live_ids = provider_model_ids(current_provider)
+        live_lower = {_strip_ollama_cloud_suffix(m.lower()) for m in live_ids}
+        if name.lower() in live_lower:
+            return None
+    except Exception:
+        pass
+
     # --- Step 2: check OpenRouter catalog ---
     # First try exact match (handles provider/model format)
     or_slug = _find_openrouter_slug(name)


andrexibiza (commented):
This is aimed at a real, expensive defect, and the placement is directionally right: ACP still calls `detect_provider_for_model()` after `parse_model_input()` when the parsed provider equals the current provider, so a bare model that overlaps OpenRouter can cross a provider/billing boundary. I traced this exact head (`53d6caabb8a6193ceb72c2ce58133e3ffb45d287`) against current `main@3340bbbdad8368e7f3d9f6827d61692adbbd87d6`, the ACP call site, `provider_model_ids()`, the Ollama Cloud discovery/cache path, current tests, exact-head workflows, and the existing model-routing/catalog graph. I do not think this head is safe to land yet, for two coupled correctness reasons.

### P1 — `provider_model_ids()` is not an authoritative live capability proof

The new Step 1.5 says the current provider's **live** catalog is authoritative, but the function it consumes has deliberately broader semantics. `hermes_cli/models.py::provider_model_ids()` returns the "best known model catalog" and, for `ollama-cloud`, delegates to `fetch_ollama_cloud_models()`. On current main that function still unions models.dev-only IDs into the live `/v1/models` result.

That is not a hypothetical edge. Open issue #94041 is the exact counterexample: the live Ollama Cloud API reported 19 models while the merged cache exposed 23; models.dev-only retired IDs such as `kimi-k2.5`, `deepseek-v4-pro`, `deepseek-v4-flash`, and `minimax-m2.5` were re-added after refresh and then failed at runtime because Ollama no longer served them. Two open repair candidates already own that adjacent work: fangliquanflq's #94055 and liuhao1024's #94062. They differ on the empty-live-result semantics, but both exist because the current union cannot be used as proof of provider existence.

That makes this change capable of producing the inverse of the incident it fixes:

`live Ollama catalog does NOT contain X` → `models.dev still contains X` → `provider_model_ids("ollama-cloud")` returns X → Step 1.5 returns `None` → Hermes falsely treats the current provider as the owner and suppresses provider resolution.

Please do not bind provider authority to this mixed-provenance list. Either land/compose the #94041 authoritative-source correction first, or introduce one typed capability seam whose result distinguishes **authoritative present / authoritative absent / unknown**. Then add the adversarial regression that matters here: live `/v1/models` excludes `X`, models.dev includes stale `X`, and `detect_provider_for_model(X, "ollama-cloud")` must not interpret the registry-only row as current-provider ownership.

Merge-order-wise, #97487 is complementary to #94041/#94055/#94062, not a replacement for them, but it currently depends on the very catalog semantics they are repairing.

### P1 — lookup uncertainty still falls through to the original paid-provider mutation

The new guard is explicitly:

```python
try:
    live_ids = provider_model_ids(current_provider)
    ...
except Exception:
    pass

# Step 2: OpenRouter fallback
```

So if current-provider capability discovery is indeterminate, the code deliberately resumes the exact automatic OpenRouter path that produced the reported charge. `provider_model_ids()` also has internal fallback behavior for several providers, so today "could not prove current provider serves this model" can collapse either to a registry-derived positive or to the old cross-provider fallback. Those are different facts and should not share one `list[str]`/exception channel.

For a path that can silently rebuild the session onto a differently billed provider, **unknown is not evidence of absence**. The routing decision needs a positive authority boundary: current-provider presence may keep the route; an authoritative absence may permit normal inference according to the product policy; an unavailable/ambiguous catalog must not silently authorize a different paid provider. Please pin that with a failure witness as well: make current-provider discovery fail while OpenRouter contains the bare name, and assert that the failure itself is not permission to mutate provider/billing authority.

This is the same defect family as the older ACP explicit-provider work (#59092/#59197/#59294): those PRs found that `detect_provider_for_model()` can overwrite already-resolved provider intent merely because OpenRouter knows the bare model. It is also adjacent to the custom-provider catalog-priority guards (#39803/#90539) and the broader silent-paid-fallback class in #81952. The common invariant is stronger than "check one more catalog before OpenRouter": a provider transition needs evidence that is scoped to the route being changed.

### Verification / landing gates

There is currently no automated regression in this PR; the only changed production file is `hermes_cli/models.py`. For a billing/routing bug, the positive case alone is not enough. I would want at least the current-provider-hit case, the #94041 stale-registry counterexample, the discovery-unknown case, and a genuine-remap control so the intended bare-model convenience remains pinned rather than accidentally disabled.

Exact-head hosted acceptance is also absent at this SHA: CI run https://github.com/NousResearch/hermes-agent/actions/runs/33217427591, Nix https://github.com/NousResearch/hermes-agent/actions/runs/33217427020, and Docker https://github.com/NousResearch/hermes-agent/actions/runs/33217427049 all completed `action_required`; the CI run has zero jobs. There is therefore no exact-head green receipt yet.

One architecture note is also relevant before this grows further: `hermes_cli/models.py` is already a tracked 2K-law fracture under #79909 / #78647. The provider-capability proof introduced here is a good bounded ownership seam; it should not become another network/cache/authority branch accreted into the godfile. Keep `detect_provider_for_model()` thin and give authoritative capability resolution one owner that #94041 and this fix can both consume.

The incident and the core insight are valid: a currently served model must not be hijacked by an aggregator just because the aggregator also knows its bare name. The part that needs tightening is the evidence used to say "currently served." Once that evidence is authoritative, preserves `unknown`, and is covered on both sides of the shape, this becomes a strong class-level repair rather than swapping a false negative for a false positive.

Adolanium:
# PR Cost & Regression Review

**PR:** [fix(provider): consult current provider's live catalog before OpenRouter fallback in detect_provider_for_model](https://github.com/NousResearch/hermes-agent/pull/97487)
**Author:** sj-unit72
**Base → Head:** main → fix/ollama-cloud-openrouter-hijack-detect
**Files changed:** 1 (net +15 / −0)
**Review date:** 2026-09-10

## 1. Summary of the Change

The PR claims that on the ACP path, a bare model name served by the current provider (e.g. `glm-5.3-flash` on `ollama-cloud`) gets silently rebuilt on OpenRouter because `detect_provider_for_model()` only consults the static catalog before falling through to the OpenRouter slug lookup. The diff adds a "Step 1.5" between the static check and the OpenRouter fallback: it calls `provider_model_ids(current_provider)`, normalizes with `_strip_ollama_cloud_suffix`, and returns `None` (stay put) on a match, wrapped in `try/except: pass` so failures keep the old behavior. The implementation matches the stated intent for the bare-name plus fresh-catalog case, but the body's "fixes the whole class" claim overreaches (see section 6).

## 2. Duplicate / Overlapping PR Check

No duplicate found. No other open, merged, or recently closed PR adds a live-catalog stay-check inside `detect_provider_for_model`. The existing reviewer already established the relationship to the ollama-cloud catalog-authority work (open issue #94041 and open repair candidates #94055, #94062, all still OPEN, complementary dependency), so I do not re-argue that here.

Novel matches from my own search:

- **#48744 - fix(cli): check configured provider catalog before static detection in /model switch** - Open
  - **Relationship:** Related but distinct
  - **Overlap:** Shared technique (consult a live catalog before static-catalog routing), but different file (`hermes_cli/model_switch.py`, Step d2), different provider notion (configured default provider, switching *to* it), different trigger (`/model` bare name). This PR stays on the *current* provider inside `detect_provider_for_model`.
  - **Impact on this PR:** Complementary, no conflict. Both can land.
- **#59294 - fix(acp): skip detect_provider_for_model for explicit provider input** - Open
  - **Relationship:** Related but distinct
  - **Overlap:** Same ACP call site (`acp_adapter/server.py::_resolve_model_selection`) and same symptom (OpenRouter hijack), but for explicit provider-prefixed input, fixed by skipping detection entirely. Prefixed inputs also bypass this PR's Step 1.5 matching (bare live ids never equal a `vendor/model` string), so the two do not interfere.
  - **Impact on this PR:** Complementary, no conflict.
- Screened and set aside as same-subsystem but distinct problems: #19950 (active-provider preference for Anthropic sonnet), #57176 (authenticated-provider preference), #56489 (credential-gated MiMo stay), #39803 (custom catalog priority), #105604 (fresh-session custom routing).

## 3. Cost Impact

### Low

- **[Latency] - Synchronous catalog fetch added to a previously pure routing function**
  - **File:** `hermes_cli/models.py`
  - **Changed lines:** approximately the new Step 1.5 block
  - **Changed hunk:**

    ```diff
    +    try:
    +        live_ids = provider_model_ids(current_provider)
    +        live_lower = {_strip_ollama_cloud_suffix(m.lower()) for m in live_ids}
    +        if name.lower() in live_lower:
    +            return None
    +    except Exception:
    +        pass
    ```

  - **Before:** `detect_provider_for_model()` performed only in-memory static lookups (static table, `_find_openrouter_slug` over a cached id list). Zero I/O.
  - **After:** every miss of the static check calls `provider_model_ids(current_provider)`, which for `ollama-cloud` resolves through `fetch_ollama_cloud_models()` (1h disk cache, else live `/v1/models` with an 8s timeout plus a models.dev query) and for other providers through its own live/profile fetchers.
  - **Cost:** Blocking network I/O on the model-resolution path.
  - **When paid:** Cache hit (the common case) costs one small file read. The full cost is paid on cold cache, first call per provider, and cache expiry.
  - **Magnitude:** Negligible in the common cached case, up to seconds on the cold path. Bounded in practice because all four call sites (`acp_adapter/server.py:317`, `hermes_cli/oneshot.py:327`, `hermes_cli/model_switch.py:1230`, `hermes_cli/web_server_config.py:765`) are model-selection events (session start, switch, config change), not per-turn hot paths.

## 4. Regression Risks

### Note

- **[Functional] - Ollama-specific suffix strip applied to every provider's catalog can falsely pin a session**
  - **File:** `hermes_cli/models.py`
  - **Changed lines:** approximately the `live_lower` line of Step 1.5
  - **Changed hunk:** same Step 1.5 hunk as above.
  - **Before:** A request for model `foo` on provider P (which serves literally `foo-cloud` but not `foo`) fell through to OpenRouter resolution.
  - **After:** The live id `foo-cloud` is stripped to `foo`, matches, and returns `None` (stay), so the session stays on P and fails at runtime with model-not-found instead of routing onward.
  - **Failure scenario:**
    1. Provider P (not ollama-cloud) serves a literal `X-cloud` / `X:cloud` id but not bare `X`.
    2. User selects bare `X` while on P.
    3. Step 1.5 strips the live id to `X`, matches, returns `None`.
    4. The session stays on P and the subsequent API call 404s instead of routing to the provider that serves `X`.
  - **Impact:** Requires a provider serving a literal `-cloud`-suffixed id, which outside the Ollama convention I could not confirm exists, so this stays a Note. The failure is also loud (API error), not silent billing.

> No regressions identified above Note severity. I explicitly checked the `current_provider == "openrouter"` path for a behavior change and found none: OpenRouter live ids are full slugs, so a bare name never matches Step 1.5 and exact slugs already returned `None` before.

One significant thing I checked and **dismissed**: at first glance `_strip_ollama_cloud_suffix` looks unimported in `hermes_cli/models.py` on current main, which would make the new block raise `NameError` (swallowed by its own `except`, silently disabling the fix). I verified against the PR's actual base (`3340bbbd`): the helper is defined directly in the pre-decomposition `models.py` godfile there, so the code runs as written. No finding.

## 5. Test Coverage of the Diff

- **Covered:** Nothing automated. The PR changes one production file and adds no tests.
- **Missing:** Current-provider hit case, stale-registry counterexample, discovery-unknown case, genuine-remap control.
- **Regression test for original issue:** No.
- **Assessment:** The specific cases worth pinning were already requested in the existing review. The manual verification in the PR body covers only the positive case.

## 6. Claim vs Reality

- **Claim: "fixes the whole class: any currently-served model can no longer be re-routed."** Reality: it fixes the bare-name plus fresh-catalog subset. A vendor-prefixed input such as `zai/glm-5.3-flash` never matches the bare live ids in Step 1.5 (`"zai/glm-5.3-flash" != "glm-5.3-flash"`), so it falls through to the exact same OpenRouter hijack for prefixed spellings. The code contains no prefix normalization on the `name` side.
- **Claim: "Wrapped in try/except so a live-fetch failure degrades to the old fallback."** Reality: accurate as a description, but the old fallback is the reported billing bug itself, so the failure mode preserves the incident rather than containing it (consequence already developed in the existing review, noted here only for completeness).
- **Claim: live incident with ~$0.29 estimated / $0.875 billed.** Author-reported, not verifiable from the diff. No reason to doubt it, recorded as unverified.

## 7. Existing Review Coverage

One substantive review (andrexibiza, COMMENTED) already clearly identified: the mixed-provenance catalog problem (Step 1.5 binds authority to `provider_model_ids()`, which unions models.dev ids per #94041, enabling the inverse false-stay), the unknown-catalog falls-through-to-paid-mutation problem, the missing automated regression coverage plus absent green CI receipt, and the godfile-seam architecture note with the related-PR family. All findings above were filtered against that review; section 4 and the #48744/#59294 entries are novel.

## 8. Overall Recommendation

**Request changes**

The incident is real and the placement is right, but this head should not land as-is. The required changes, in severity order:

1. Do not bind the stay decision to the mixed-provenance catalog (existing P1, already raised in review): either land the #94041 authority correction first or gate Step 1.5 on an explicitly authoritative presence signal, with the stale-registry adversarial test.
2. Do not let an indeterminate catalog fall through to the paid OpenRouter mutation (existing P1, already raised in review): unknown must preserve the current route or ask, with a failure-witness test.
3. Rebase: the branch reports `CONFLICTING` against current main (no CI checks reported), so the hunk placement must be re-confirmed after rebasing past the models decomposition.
4. Narrow the residual gap behind the "whole class" claim: normalize the vendor prefix on the `name` side (or scope the claim to bare names), so prefixed spellings get the same protection.

## 9. Unverified / Needs Author Clarification

- The billed-amount figures in the PR body are author-reported and cannot be confirmed from the diff.
- Whether any non-Ollama provider serves a literal `-cloud`-suffixed id (needed to promote the section 4 Note): a negative answer would dismiss it entirely.
- The duplicate search covered open/closed PRs by keyword, function name, and symptom; a differently-worded fix for the same ACP hijack could still exist outside those signals.

teknium1:
Thanks @sj-unit72 — you diagnosed this class correctly and put the guard at the right function. The ollama-cloud `glm-5.3-flash` case, plus the Codex `gpt-6-astra` and Nous-sub variants that surfaced this week, are now fixed on main via https://github.com/NousResearch/hermes-agent/pull/107236 (`215fd0ecb9a`).

What landed differs from this diff in three ways, which is why it was reimplemented rather than cherry-picked: it reads the current provider's **disk-cached** catalog (`cached_provider_model_ids`, 1h TTL + stale-while-revalidate) instead of a blocking `provider_model_ids()` fetch on the switch path; it resolves a bare name to the provider's own full slug (`glm-5.3-flash` → `zai/glm-5.3-flash` on Nous) instead of only returning "stay"; and it drops the ollama-only suffix strip that would have applied to every provider. Your PR is credited in the merge and commit body. Closing as superseded.

PR #76850, opened 2026-08-02
fix(buzz): handle restricted CLOSED per-subscription, stop reconnect flood

xozai:
## What does this PR do?

When a Buzz relay sends a `CLOSED` frame for a single subscription with a `restricted: not a channel member` error, the adapter was raising `ConnectionError`, tearing down the entire WebSocket connection, and immediately reconnecting — causing a ~1.6 s reconnect flood in `gateway.log`.

**Root cause:** The `CLOSED` handler in `_websocket_loop` unconditionally raised `ConnectionError` regardless of whether the error was permanent (restricted) or transient (server shutdown).

**Fix:**
- On a `restricted` CLOSED, drop only the offending subscription and record the channel in a new `_restricted_channels` set instead of tearing down the whole WS connection.
- Skip channels in `_restricted_channels` during `connect()` seeding and `_subscribe_websocket()` so reconnects don't re-trigger the same rejection.
- Non-restricted CLOSED frames still raise `ConnectionError` and reconnect as before.

## Related Issue

Fixes # (no existing issue — discovered in production on a relay with private channels)

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✅ Tests (adding or improving test coverage)

## Changes Made

- `plugins/platforms/buzz/adapter.py`: add `_restricted_channels` set; handle `CLOSED` per-subscription; skip restricted channels in `connect()` and `_subscribe_websocket()`
- `tests/gateway/test_buzz_websocket.py`: three new regression tests

## How to Test

1. Configure the Buzz adapter against a relay where the agent's identity is a member of some channels but not others (e.g. private channels it hasn't been invited to)
2. Start the gateway — before this fix, `gateway.log` floods with `WebSocket disconnected; retrying in 1.0s: restricted: not a channel member` every ~1.6 s
3. After this fix: a single WARNING is logged per restricted channel (`relay permanently rejected channel <id>`) and the connection stays stable

### Code

- [x] I've read the Contributing Guide
- [x] My commit messages follow Conventional Commits
- [x] I searched for existing PRs — no duplicates found
- [x] My PR contains only changes related to this fix
- [x] I've run `pytest tests/gateway/test_buzz_websocket.py tests/gateway/test_buzz_adapter.py -q` — 30 passed
- [x] I've added tests for my changes
- [x] Tested on: macOS 15 (Apple Silicon) against buzz.xozai.com

### Documentation & Housekeeping

- [x] N/A — no config keys added, no docs changes needed

[...]

PR #97502, opened 2026-08-28
fix(buzz): enforce member-only subscriptions instead of reconnect-looping

repfigit:
## Summary

A channel the gateway key is not a member of produced a `CLOSED` frame (`restricted: not a channel member`); the Buzz adapter treated **any** `CLOSED` frame as a transport error and reconnect-looped at 1s intervals. In production this produced 13.6k log lines in a single day, with the relay seeing the same churn. Retrying a membership rejection can never succeed.

## Changes

Two-layer fix in `plugins/platforms/buzz/adapter.py`:

1. **Connect-time membership pre-filter** — before subscribing to anything, each candidate channel gets a real `messages get --limit 1` probe. A channel the key cannot read is dropped from the watch set with a warning, so a membership rejection never reaches the WebSocket transport.

2. **Runtime `CLOSED` handling** — a `CLOSED` frame naming a membership restriction (`not a channel member`, `restricted`, `auth-required`) now prunes that channel from the watch state and keeps the socket alive. Only genuine transport errors still trigger reconnect. Previously the pruning branch did not exist: every `CLOSED` raised `ConnectionError`, tore down the socket, re-authenticated, re-subscribed to the same rejected channel, and repeated forever.

3. **Send-failure workaround (bonus, same file)** — `buzz messages send` hard-fails on any `@word` token that does not resolve to a channel member, which silently drops agent replies containing `@session:...` links or prose like `@someone`. `send_message` now retries once with `@` escaped to full-width `＠` (presentation-only) so the reply still lands. [Drafted upstream for the CLI side](https://github.com/block/buzz).

## Test plan

- [x] New `tests/gateway/test_buzz_membership_filter.py`: pre-filter drops non-member channels (stubbed CLI), `CLOSED`-branch ordering in `_websocket_loop` source, and the state transition (channel removed from `_channel_state`/`_channel_names`/subscriptions, socket survives)
- [x] Full Buzz suite green: 28 passed, 0 failed with this change (2 pre-existing failures on `main` are unrelated — they also fail on a clean checkout of `main` under this environment's pytest invocation)
- [x] `ruff check` clean on both files
- [x] Validated against a live self-hosted relay: after restart the gateway watches 6 channels, all member-readable, zero membership errors in the log

### DIFF
--- a/plugins/platforms/buzz/adapter.py
+++ b/plugins/platforms/buzz/adapter.py
@@ -531,6 +531,27 @@ class BuzzAdapter(BasePlatformAdapter):
             logger.error("Buzz: no channels to watch (configure BUZZ_CHANNELS or join a channel)")
             self._set_fatal_error("config_missing", "no Buzz channels to watch", retryable=False)
             return False
+        # Membership filter: only watch channels this key can actually read.
+        # An explicit channels: entry the key is not a member of would
+        # otherwise CLOSED-loop at reconnect time (retries can never succeed).
+        member_channels: List[str] = []
+        for channel_id in watch:
+            code, _out, probe_err = await self._run_cli(
+                ["messages", "get", "--channel", channel_id, "--limit", "1"]
+            )
+            if code == 0 or "not a channel member" not in (probe_err or ""):
+                member_channels.append(channel_id)
+            else:
+                logger.warning(
+                    "Buzz: skipping channel %s — key is not a member (%s)",
+                    self._channel_names.get(channel_id, channel_id),
+                    probe_err,
+                )
+        watch = member_channels
+        if not watch:
+            logger.error("Buzz: no watchable channels remain after membership filter")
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: #76850 fixes the same membership-rejection reconnect loop with a persistent rejected-channel set. This PR adds connect-time membership filtering plus an outbound @-token fallback; maintainers should compare and consolidate the overlapping adapter changes.

teknium1:
Closing with credit — your runtime CLOSED matching phrases ('not a channel member', 'auth-required') were composed into #76850's handler via #99430 with credit in the commit message. The per-channel connect-time probe was declined (N extra CLI calls per connect, redundant once runtime pruning + the persistent restricted set exist) and the @-escape send fallback landed separately via the dispatch cluster (#99431). Thanks for the sharp diagnosis of the reconnect loop.

Issue #95550, opened 2026-08-26
fix(cli): Shift+key sequences (ESC[27;2;<cp>~) leak as literal text into the prompt on Ghostty via SSH since the Ghostty modifyOtherKeys-only path


juty9026:
### Environment
- Hermes Agent: v0.20.5 (2026.8.19), git install (`d22e2b9f6e`)
- Terminal: Ghostty (macOS), `TERM=xterm-ghostty`
- Connection: `ssh <vps>` → Ubuntu 6.17 kernel → hermes classic CLI (prompt_toolkit 3.0.52)
- Also reproducible inside tmux on the VPS

### Summary

Since the Ghostty exception path was introduced (skip kitty protocol, push modifyOtherKeys only — `1a8fea3ce2`, shipped around v2026.8.16+), Shift+key presses over Ghostty are inserted into the prompt as literal text:

- Shift+A        → `^[[27;2;65~`
- Shift+backtick → `[27;2;126~`

This worked before v2026.8.16-era builds. Workaround is `hermes config set display.cli_multiline_shortcuts false`, which stops the extended-key push entirely but also disables Shift/Ctrl+Enter newline.

### Root cause analysis

`_enable_extended_enter_keys()` (cli.py:4279) pushes `ESC[>4;2m` for Ghostty (`_is_ghostty_terminal`). Under modifyOtherKeys=2 Ghostty re-encodes every Shift combo as `ESC[27;<mod>;<codepoint>~`.

Two gaps:

**1. Mapped sequences still leak at runtime.**
`install_modify_other_keys_aliases()` (hermes_cli/pt_input_extras.py) *does* install `"\x1b[27;2;65~" -> "A"`, and a standalone Vt100Parser simulation converts it correctly:

```python
from prompt_toolkit.input.vt100_parser import Vt100Parser
out = []
p = Vt100Parser(out.append)
for ch in "\x1b[27;2;65~": p.feed(ch)
p.flush()
# -> KeyPress(key='A')
```

But in a real running CLI the same bytes land in the buffer as literal text. Reproduced headlessly on the VPS:

```bash
tmux new-session -d -s t -x 120 -y 40 'venv/bin/hermes'
sleep 8
tmux send-keys -t t -l $'\x1b[27;2;65~'
tmux capture-pane -t t -p | grep '❯'
# ❯ ^[[27;2;65~
```

[...]

Issue #93633, opened 2026-08-24
[Bug]: Classic CLI prints literal [27;2;64~ for Shift+symbol keys (@ ~ ! $ ...) under modifyOtherKeys=2 on Ghostty — #87511 covered letters only


turingcat:
## Summary

In the classic interactive CLI, **Shift + symbol keys** (`@ ~ ! $ % ^ & * ( ) _ + { } | : " < > ?`) print literal text like `[27;2;64~` into the prompt instead of the symbol, on terminals that emit **xterm modifyOtherKeys level 2** (Ghostty, VS Code terminal, iTerm2).

`#87511` fixed Shift+letter (`ESC[27;2;97~` → `A`), and its PR deliberately left symbols unmapped with the note *"Shift+digit symbols are layout-specific … they will leak, but that's better than wrong input."* That caution is right for **base-digit** codepoints, but wrong for the codepoints these terminals actually send.

## Environment

- Hermes Agent v0.20.4 (2026.8.22), git install, current `main` (7e67f64fce)
- Python 3.14, prompt_toolkit 3.0.52
- macOS (Apple Silicon)
- Terminal: cmux embedding Ghostty 1.3.2 → `TERM_PROGRAM=ghostty`, `TERM=xterm-256color` (cmux does not set `TERM=xterm-ghostty`), Hermes detected Ghostty and enabled extended keys via the `TERM_PROGRAM` allowlist

## Steps to reproduce

1. Start the classic CLI (`hermes`) in Ghostty (directly or inside cmux).
2. Type `Shift+2` (US layout: `@`).
3. Observe the prompt.

## Expected vs actual

- Expected: `@` inserted.
- Actual: `[27;2;64~` inserted literally (ESC consumed by the parser; the rest lands in the buffer). Same for `~ ! $ & _` etc. — **every Shift+symbol**.

## Root cause

Hermes pushes `CSI >4;2m` (modifyOtherKeys=2) to allowlisted terminals. Under level 2, Ghostty/VS Code/iTerm2 re-encode modified keys as `CSI 27;<mod>;<codepoint>~`, and for Shift+symbol they report the **shifted** codepoint: `Shift+2` → `ESC[27;2;64~` where **64 is `@` itself**.

`install_modify_other_keys_aliases()` in `hermes_cli/pt_input_extras.py` maps Shift+letter but skips symbols, so these sequences fall through the VT100 parser and leak as literal text.

## Why self-mapping the codepoint is layout-safe

[...]

PR #97503, opened 2026-08-28
fix(cli,tui): Shift+key types the character under Ghostty modifyOtherKeys

MKappes:
### Environment

- Terminal: Ghostty 1.3.1 (macOS), `TERM=xterm-ghostty`, `TERM_PROGRAM=ghostty`
- Hermes Agent v0.20.6 (2026.8.27)
- Affects both the classic CLI and the Ink TUI

### Summary

Both surfaces push xterm `modifyOtherKeys` level 2 (`CSI >4;2m`) for Ghostty, so the terminal re-encodes every Shift combo as `ESC[27;<mod>;<codepoint>~`. Three separate defects made those keypresses produce the wrong text. Typing a capital letter in the classic CLI inserted `^[[27;2;84~`; in the Ink TUI it inserted a lowercase `t`.

### Root cause

**1. Classic CLI — wrong data payload (the divergence #95550 flagged as needing a closer look).**

`install_modify_other_keys_aliases()` resolves the sequence correctly:

```python
ANSI_SEQUENCES['\x1b[27;2;84~']  # -> 'T'
```

But `Vt100Parser._call_handler` builds `KeyPress(key, data)` where `data` is the *matched byte sequence*, and prompt_toolkit's default `Keys.Any` binding inserts `event.data` (`key_binding/bindings/basic.py`). So the parser found the right key and typed the raw escape:

```
KeyPress(key='T', data='\x1b[27;2;84~')
```

This is exactly why the standalone `Vt100Parser` simulation in #95550 reported `key='A'` and looked correct while the live CLI still leaked. The alias table was never the problem; the payload was.

`install_plain_char_data_fix()` wraps `_call_handler` so a single-character key whose data is an escape sequence carries the character as its data. `Keys.*` payloads are left alone, so `Ctrl+C`, arrows, and the `(Escape, ControlM)` Shift+Enter tuple keep their raw data.

**2. Classic CLI — shifted symbols had no mapping at all.**

`shift_map` covered letters only, with an explicit comment that shifted digits are layout-specific (US `!` vs AZERTY `¹`). That reasoning holds for the *unshifted* codepoint, but modifyOtherKeys emitters report the **already-shifted** codepoint — Ghostty sends `ESC[27;2;126~` for Shift+`` ` ``. The codepoint *is* the character, so no layout knowledge is required. Printable ASCII 33–126 now maps to itself via `setdefault`, leaving the letter entries and any earlier registration to win.

**3. Ink TUI — shifted letters arrived lowercase (not previously reported).**

No escape leak here: `input-event.ts` already extracts `[27;…~` sequences. The bug is that `keycodeToName()` calls `.toLowerCase()` on every printable codepoint so binding lookups have a stable name, and `parseKey()` hands that lowercased name straight to text input:

```
Shift+T -> input="t" shift=true
```

Under Ghostty's modifyOtherKeys push the TUI produced **no capital letters at all**. `parseKey()` now re-uppercases a single a–z character when the keypress came through a special sequence with plain Shift and no ctrl/meta/super/option. The modifier guard matters: with those set, `input` is a binding discriminator that consumers match lowercase, so `Ctrl+C` and `Ctrl+Shift+C` stay `c`.

### Verification

Verified end to end in both live applications, not only in a parser simulation — given how #95550 played out, that distinction seemed worth insisting on. tmux, `TERM=xterm-ghostty`, feeding the raw bytes for Shift+T, H, I, S, `~` and `!`:

```bash
tmux new-session -d -s t -x 120 -y 40 "TERM=xterm-ghostty TERM_PROGRAM=ghostty ./venv/bin/hermes"
tmux send-keys -t t -l $'\x1b[27;2;84~\x1b[27;2;72~\x1b[27;2;73~\x1b[27;2;83~\x1b[27;2;126~\x1b[27;2;33~'
tmux capture-pane -t t -p | tail -1
```

| Surface | Before | After |
|---|---|---|
| Classic CLI | `^[[27;2;84~` | `Ψ THIS~!` |
| Ink TUI | `this` | `Ψ THIS~!` |

### Tests

- `tests/cli/test_shift_modify_other_keys.py` — both encodings (`ESC[27;2;CP~` and `ESC[CP;2u`), shifted symbols, untouched control keys, idempotent install, and a loop over codepoints 33–126 asserting no `ESC` or `[27;` reaches the buffer.
- `ui-tui/packages/hermes-ink/src/ink/events/shifted-letters.test.ts` — the TUI equivalent, including a Ctrl+Shift case that pins the lowercase discriminator so the guard can't be dropped later.

Suites: Python CLI key handling `243 passed, 2 skipped`; full ui-tui suite `160 files, 1720 passed, 3 skipped, 0 failed`.

### Relationship to existing work

Refs #95550. Overlaps PR #95562, which addresses the CLI plain-char case only — it does not cover the shifted-symbol gap (#2) or the TUI casing bug (#3). Happy to rebase onto that PR if maintainers prefer to land it first; the shifted-symbol and TUI changes are independent of the mechanism chosen for #1.

### One design question for reviewers

Fix #1 monkeypatches a prompt_toolkit internal (`Vt100Parser._call_handler`). It is guarded and idempotent, but a prompt_toolkit major bump could move that method. The alternative is fixing the `Keys.Any` binding to prefer `event.key` for single-character keys, which is arguably where the bug lives but has a wider blast radius across every key path. I went with the narrower change; happy to switch if you'd rather own it at the binding layer.

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -82,6 +82,7 @@ try:
         install_ctrl_enter_alias,
         install_ignored_terminal_sequences,
         install_modify_other_keys_aliases,
+        install_plain_char_data_fix,
         install_shift_enter_alias,
     )
     install_shift_enter_alias()
@@ -89,7 +90,8 @@ try:
     install_cmd_backspace_alias()
     install_modify_other_keys_aliases()
     install_ignored_terminal_sequences()
-    del install_shift_enter_alias, install_ctrl_enter_alias, install_cmd_backspace_alias, install_modify_other_keys_aliases, install_ignored_terminal_sequences
+    install_plain_char_data_fix()
+    del install_shift_enter_alias, install_ctrl_enter_alias, install_cmd_backspace_alias, install_modify_other_keys_aliases, install_ignored_terminal_sequences, install_plain_char_data_fix
 except Exception:
     pass
 import threading
--- a/hermes_cli/pt_input_extras.py
+++ b/hermes_cli/pt_input_extras.py
@@ -331,6 +331,15 @@ def install_modify_other_keys_aliases() -> int:
         upper_char = chr(ch - 32)  # 'A'..'Z'
         shift_map[ch] = upper_char
         shift_map[ch - 32] = upper_char
+    # Shift+symbol: emitters that report the ALREADY-SHIFTED codepoint (the
+    # xterm modifyOtherKeys behaviour — Ghostty sends ESC[27;2;126~ for
+    # Shift+`) need no layout knowledge at all: the codepoint IS the
+    # character to insert.  Map every remaining printable ASCII codepoint to
+    # itself so '~', '!', '@', '{' etc. stop leaking (#95550).  Unshifted
+    # codepoints are deliberately still NOT guessed at, since the shifted
+    # result of a digit/punctuation key is layout-specific.
+    for cp in range(33, 127):
+        shift_map.setdefault(cp, chr(cp))
     _install_paired(2, shift_map)
 
     # -- Multi-modifier letters: Shift+Alt (4), Ctrl+Shift (6),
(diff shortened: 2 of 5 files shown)


kvnloo (commented):
Verified at exact head 949987845ef3a933f13e1242ab255fed541a1d44. The prompt_toolkit data rewrite is restricted to resolved single-character keys with a distinct ESC-prefixed payload, and the Ink casing repair is restricted to plain-Shift special-sequence ASCII letters; control and multi-modifier paths remain unchanged. Independent focused runs passed (Python: 225 key-alias tests, plus 19 passed/1 environment-dependent skip across Ctrl/Shift+Enter suites; Ink: 7 tests; TypeScript typecheck). I found no blocking correctness, security, compatibility, privacy, or regression issue. Approve.

braden-dev:
Independent confirmation of the root-cause analysis here, from a separate debugging effort on Linux (Fedora 44, Ghostty, TERM=xterm-ghostty, Hermes 0.20.6, commit ac6c8028e).

We hit the same symptom cluster (Shift+letter typing `^[[27;2;<cp>~` in the classic CLI, lowercase letters in the TUI) and root-caused it independently to the same three defects before finding this PR:

1. `Vt100Parser._call_handler` builds `KeyPress(key, data)` with the raw matched sequence as `data`, and the `Keys.Any` self-insert binding inserts `event.data` verbatim — so a correctly-decoded `'A'` still types the raw escape. We can confirm the misleading part: a standalone `Vt100Parser` simulation shows `key='A'` and looks fixed while the live app still leaks. Only patching the data payload fixed it.
2. Shift+symbol codepoints (`ESC[27;2;64~` = `@` etc.) additionally have no ANSI_SEQUENCES mapping at all (the #93633 half).
3. The Ink TUI's `keycodeToName()` lowercases codepoints, destroying case in the composer.

Verification method that finally reproduced it reliably, in case it helps review: run `hermes --cli` inside tmux and inject the exact bytes with `tmux send-keys -H 1b 5b 32 37 3b 32 3b 36 35 7e ...` — before the data-payload fix the prompt shows the raw sequences; after it, `ASDF@\"` lands correctly, normal typing and Ctrl+A/Ctrl+C bindings unaffected. An equivalent fix has been running locally on our machine since last night with no regressions in daily use.

PR #95899, opened 2026-08-26
fix(cron): multiplex deliveries use the owning profile's adapters and credentials

jackthelobster:
## Summary

Under a multiplexed gateway (`gateway.multiplex_profiles: true`), scheduled cron jobs belonging to secondary profiles **execute correctly but deliver through the wrong bot identity**. The in-process cron ticker only ever receives `runner.adapters` — the *active* profile's flat platform map — so every secondary profile's delivery resolves whatever other identity holds the platform slot. On Discord this means sending a profile's DM brief as a bot that is not a participant of that conversation; Discord rejects it with `403 Forbidden (error code: 50001): Missing Access`. The job reports `last_status: ok` with an error string in `last_delivery_error`, and the output is silently stranded.

The standalone delivery fallback cannot rescue it either: `run_one_job` resets the job's secret scope before `_deliver_result` runs (correctly — agent-turn credentials must not leak past the turn), so the fallback resolves platform tokens from process-global `os.environ`, which under a multiplexer holds whichever profile started the process. Again the wrong bot.

This PR fixes both halves so delivery identity always matches the owning profile:

- **`gateway/run.py`** composes a `profile_adapters` map (profile name → that profile's live adapters). The active profile keeps `runner.adapters`; secondaries resolve their entry from `runner._profile_adapters`; an unconnected/disabled secondary gets `{}` rather than another profile's map. Maps are shared **by reference**, so reconnect/swap bookkeeping performed on the runner stays visible to the ticker.
- **`cron/scheduler_provider.py`** (`InProcessCronScheduler.start` / `_start_multiplex`) hands each profile's tick ONLY its own adapter map via the new `cron.scheduler._adapters_for_profile`. A profile with no connected adapters resolves no live transport at all — never another profile's bot.
- **`cron/scheduler.py`** adds `_delivery_secret_scope` and wraps both `_deliver_result` call sites (normal delivery and failure-alert delivery) so the standalone fallback reads the owning profile's `<home>/.env` plus its configured external secret sources, mirroring exactly what `run_job` already does per job run. Gated on multiplexing being active, so single-profile processes keep legacy behavior byte-for-byte.

## Reproduction

1. Run two profiles under one multiplexer, each with its own Discord bot token (e.g. default + coach).
2. From the coach bot's DM with user U, schedule any cron job with `deliver` targeting that DM.
3. At fire time the job runs and saves output, but delivery fails: `403 Forbidden (50001): Missing Access`, emitted by the **default profile's** discord adapter logger, every night.
[...]

PR #91953, opened 2026-08-22
fix(cron): gate adapter-less tickers on gateway liveness

aryn-lacy:
## Problem

When the desktop app (or any non-gateway process) runs the cron ticker, it fires jobs with `adapters=None`. Matrix delivery then falls through to the standalone path, which boots a *second* OlmMachine against the same `crypto.db` the gateway owns. Two E2EE state machines on one store corrupt megolm session state: devices stop receiving keys (UTD), one-time-key pools are silently consumed.

Not theoretical — a full production incident on 2026-08-21 traced to exactly this race (three ephemeral boots in one day exhausted the device OTK pool).

## Fix (two layers)

1. **Root cause (`web_server.py`):** the desktop cron ticker checks gateway liveness via `gateway.status.get_running_pid()` before dispatching; when a gateway owns the profile, the ticker defers (`can_dispatch` callback threaded through `InProcessCronScheduler.start()`).
2. **Blast wall (`send_message_tool.py`):** the ephemeral Matrix send path refuses to boot an OlmMachine when a gateway process is detected. Fails open if `gateway.status` cannot be imported.

## Tests

- 6 tests across `test_send_message_tool.py` / `test_scheduler_provider.py` (1 updated + 2 new per layer)
- Negative-tested: reverting source while keeping tests makes behavioral tests fail
- Full cron + tools suites green on current main (82/82 in touched files)

PR #87645, opened 2026-08-16
fix(cron): skip profile .env reloads while multiplex isolation is active

Christopher-Schulze:
## What does this PR do?

On multiplexed gateways (`gateway.multiplex_profiles: true`), `run_job`'s own `.env` reloads leak the firing profile's secrets into the shared process environment. The multiplex cron ticker already scopes each fire to the firing profile's `HERMES_HOME`, and `_run_one_job_body` installs that profile's secret and terminal scope around run and delivery — but both of `run_job`'s `load_hermes_dotenv` calls still pushed the firing profile's `.env` values into the shared process-global `os.environ`, where every OTHER profile's subprocesses inherit them after the fire.

This PR closes that remaining gap: both reload sites are skipped while profile isolation is active. The firing profile's secrets stay authoritative through the secret scope, and single-profile gateways keep the reloads unchanged (the no_agent reload that resolves auto-delivery home channels, and the pre-run reload that re-resolves Bitwarden/BSM-backed secrets, #33465).

Scope note: the PR originally also extracted `profile_runtime_scope` into `agent/secret_scope.py` and wrapped cron execution with it. Upstream `main` has since landed equivalent (and stronger) coverage natively — the multiplex ticker wraps each profile's tick in `set_hermes_home_override` + `use_cron_store`, and `_run_one_job_body` now installs the profile's secret scope plus the terminal-policy seam (#68559). Those hunks were dropped in this rebase; what remains is the piece `main` does not cover.

## Related Issue

Closes #87575

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- **`cron/scheduler.py`** (`run_job`, both reload sites):
  - `no_agent` script path: `load_hermes_dotenv(hermes_home=...)` now runs only when `not is_multiplex_active()`.
  - agent path pre-run reload: same guard, after `reset_secret_source_cache()`.
  - Both sites keep the single-profile behavior byte-identical; under multiplex isolation the reload is a no-op because the firing profile's secrets are already authoritative through the secret scope.
- **`tests/cron/test_cron_multiplex_dotenv_isolation.py`**: new regression tests pinning both sites: reload skipped under an active multiplexer, still performed for single-profile gateways. The two multiplex tests fail on `origin/main` and pass with the guard.

## How to Test

1. Run the focused regression tests:
   ```bash
   python -m pytest tests/cron/test_cron_multiplex_dotenv_isolation.py
   ```
[...]

Issue #87575, opened 2026-08-16
[Bug]: profile scope bleeds across concurrent turns — an in-process cron job for one profile makes another profile's live turn resolve HERMES_HOME to it


neo-wanderer:
### Bug Description

On a single multiplexed gateway (`gateway.multiplex_profiles: true`) serving a
default profile `alpha` and a secondary profile `beta`, an **agent-type cron job
belonging to `beta` causes a concurrent live turn of `alpha` to resolve
`HERMES_HOME` to `beta`'s home.**

The bleed lasts as long as the cron job runs and clears when it finishes, so it
is per-turn state, not a process-global environment mutation.

**Impact.** `_profile_runtime_scope` (`gateway/run.py:2014`) bundles
`set_hermes_home_override` *and* `set_secret_scope`. A bled home override means
config, skills, memory, SOUL and sessions resolve to the wrong profile for the
duration of the overlap. The secret scope is installed by the same call, so a
profile plausibly resolves another profile's credentials during that window —
I have measured only the home override, and report the credential half as
suspected rather than confirmed.

It is also visible to every shelled-out command, not just to Hermes internals:
`_inject_context_hermes_home` (`tools/environments/local.py:397`) copies
`get_hermes_home_override()` into every `terminal` child environment. Any CLI
that keys off `HERMES_HOME` — or off a variable resolved through the scope —
silently reads the wrong profile's state.

### Steps to Reproduce

1. One gateway with `gateway.multiplex_profiles: true`. Default profile
   `alpha`, secondary profile `beta`.
2. Create a cron job on `beta` whose only action is a long no-op:

   ```
   hermes -p beta cron create '*/5 * * * *' \
     'Call the terminal tool exactly once with the command `sleep 150`. When it
      returns, reply with the single line: done. Do not call any other tool.' \
     --deliver local
   ```

3. Let the **gateway's own scheduler** fire it. (Triggering it from a CLI does
   not reproduce — see "Notes" below.)
4. While it is running, on a live channel served by `alpha`, ask that agent to
   run:

   ```
   echo "HERMES_HOME=[$HERMES_HOME]"
   ```

**Expected:** `alpha`'s turn reports `alpha`'s home (`/opt/data`).

**Actual:** `/opt/data/profiles/beta`.

Reproduced on two consecutive turns 43 seconds apart inside one 150-second
window. Outside any window the same session correctly reports `/opt/data`.

### Notes on reproducing

[...]

PR #97504, opened 2026-08-28
fix(cron): stop Desktop ticks leaking profile sessions across state stores

### LINKED ISSUE #97489: Desktop cron ticker persists other profiles' cron sessions into the wrong profile's state.db
# Desktop cron ticker persists other profiles' cron sessions into the wrong profile's state.db

## Component
`hermes_cli/web_server.py` — `_start_desktop_cron_ticker` (HERMES_DESKTOP=1 path) + `cron/scheduler_provider.py` multiplex tick.

## Environment
- Hermes Agent v0.20.0 (2026.8.3), Linux, 8 profiles, each running its own per-profile gateway as a systemd user service
- Desktop app running alongside the gateways
- Built-in cron provider (`InProcessCronScheduler`), multiplex tick over 8 profile homes

## Summary
When the Hermes desktop app is running, its dashboard backend starts a desktop cron ticker that multiplex-ticks **every** profile's cron store (by design — the desktop spawns a dashboard, not a gateway). The tick loop scopes each profile via `set_hermes_home_override()` + `use_cron_store()`.

The override correctly scopes **agent execution**: a north-caribbean job fired by the desktop ticker ran with north-caribbean's system prompt, skills, config, and credentials (verified: session row `profile_name='north-caribbean'`; all 26 path references in the session point to `profiles/north-caribbean`).

However, the **session persistence layer is not swapped** during the override. The session record, messages, system-prompt copy, execution rows, and log lines were all written into the **dashboard backend's own profile home** (tgc-sales), not the job-owning profile (north-caribbean):

- `tgc-sales/state.db`: session `cron_f4202e7eef8e_20260828_180019` (58 messages, 33 tool calls, full system prompt hash `35f8c4f6…`, execution rows)
- `tgc-sales/logs/agent.log`: 56 entries for the run (2026-08-28 18:00:19–18:22:29 UTC)
- `north-caribbean/state.db`: **zero** entries for the same session — the owning profile has no record its job ran

(Reasoning: `use_cron_store(home)` swaps the cron store, and `set_hermes_home_override` swaps the agent-facing home, but the dashboard backend's already-open SQLite session DB connection and its log handlers are process-bound to the backend's startup profile and are not re-bound per tick. The output `.md` did land in the correct profile's `cron/output/` because that path is resolved inside the tick's override.)

## Repro
1. Run `hermes desktop` with 2+ profiles; ensure the desktop app's active profile is B (e.g. switch to B in the app or have it be the last active).
2. Have profile A's gateway stopped, or simply let the desktop ticker win the `.tick.lock` for a due job owned by A (tick lock is cross-process; whichever ticker grabs it first runs the job).
3. Wait for a cron job owned by profile A to fire.
4. Observe: the job executes with A's context, but the session/messages/system-prompt/logs are persisted into B's `state.db` and `logs/`.

Easiest deterministic repro: stop A's gateway so only the desktop ticker can fire A's jobs.

## Impact
- **Audit-trail breakage:** the owning profile has no session record of its own cron run; `session_search` / session history for that run lives in a different profile's database.
- **Cross-profile data mixing:** one profile's session content (prompts, tool outputs, possibly secrets visible in tool results) is persisted into another profile's stores. For users running one profile per business/client with a strict isolation requirement, this defeats the isolation guarantee at the persistence layer even though execution is scoped correctly.
- Silent — no error is logged anywhere; the job reports success.

## Expected behavior
A cron job owned by profile A should persist its session record, messages, system prompt, execution rows, and logs into profile A's stores, regardless of which process's ticker fires it. The per-tick override needs to swap (or per-tick route) the session DB and log handlers the same way it swaps the cron store and agent home — or the fire should be handed back to the owning profile's process.

## Suggested fix directions
1. Re-bind the session store + log handlers inside the multiplex tick's per-profile scope (same pattern as `use_cron_store`), so all persistence follows the override.
2. Or: have the desktop ticker only **claim** the tick and dispatch the fire to the owning profile's gateway via its control socket (the gateways are the natural execution owners when running).
3. Minimum: detect the condition and log a warning so the misplacement is visible.

## Secondary observation
Because the desktop ticker competes with all 8 per-profile gateways for each profile's `.tick.lock`, which process fires a given job is nondeterministic. This is documented as safe for double-firing (flock), but it means users running per-profile gateways get their cron jobs executed by whichever process wins — for the desktop path, with the persistence bug above. A "desktop ticker stands down for profiles that have a live gateway" rule would make ownership deterministic and eliminate this class of issue for gateway users.

## Logs/evidence (available on request)
- `~/.hermes/profiles/tgc-sales/logs/agent.log` 2026-08-28 18:00:19.596 → 18:22:29 (full run under `cron_f4202e7eef8e_20260828_180019`)
- `~/.hermes/profiles/tgc-sales/state.db` sessions/messages/system_prompts rows for that session
- `~/.hermes/profiles/north-caribbean/logs/agent.log`: zero entries in the same window
- `~/.hermes/profiles/north-caribbean/cron/output/f4202e7eef8e/2026-08-28_18-22-27.md` (output path was correct)

fangliquanflq:
## What does this PR do?

Desktop and multiplex gateway cron ticks now persist a job's session in the owning profile's `state.db`. Without this fix, the bounded SessionDB initialization worker loses the active profile ContextVar and silently writes prompts, messages, tool calls, and execution-linked session data into the ticker process profile's store.

### Symptom

When a Desktop backend or multiplex gateway fires a due job for profile A while running under profile B, the job executes with A's configuration but its cron session appears in B's session history. Profile A has no session record of its own run.

### Impact

This breaks the owning profile's audit trail and mixes potentially sensitive session content across profile storage boundaries.

### Bug Cause

**Trigger:** `cron/scheduler.py:6369` in `run_job`, when `cron.session_db_timeout_seconds` is positive (the default).

**Causal chain:**
1. The multiplex scheduler installs profile A's `HERMES_HOME` ContextVar override and dispatches the due job with a copied context.
2. `run_job` opens `SessionDB` in a nested `ThreadPoolExecutor` to bound initialization time, but that executor does not copy ContextVars.
3. `SessionDB()` resolves its default path from process profile B, and the agent persists profile A's complete cron session through that wrong handle.

**Why it is wrong:** The timeout worker crosses a thread boundary without carrying the profile identity that owns every persistence decision for the job.

**Working sibling / contrast:** The unlimited SessionDB initialization path runs inline and follows the active override. The outer cron dispatch and agent inactivity worker already use `contextvars.copy_context()` before crossing executor boundaries.

**Ruled out:** The cron store and outer job executor are already scoped correctly; the regression test observes the correct profile before the nested bounded init and the process profile only inside that worker.

### Fix

Copy the active context immediately before submitting bounded SessionDB initialization and enter it inside the worker. The existing timeout, late-result cleanup, and inline unlimited behavior remain unchanged. A regression test uses distinct process and job-owner homes and proves the constructed session handle follows the owner.

## Related Issue

Closes #97489

## Type of Change

- [x] Security fix

## Changes Made

- `cron/scheduler.py` - preserve the active profile context in the bounded SessionDB initialization worker.
- `tests/cron/test_sessiondb_init_hang.py` - reproduce distinct process/profile homes and assert the agent receives a profile-owned session store.

## How to Test

1. Run the focused cross-profile regression; it fails on current `main` by observing the dashboard home and passes with this patch by observing the job-owner home.
2. Run the related scheduler provider and SessionDB timeout suites:

```bash
scripts/run_tests.sh tests/cron/test_sessiondb_init_hang.py tests/cron/test_scheduler_provider.py -q
```

Result: 42 passed.

3. `uv run ruff check cron/scheduler.py tests/cron/test_sessiondb_init_hang.py`

Result: clean.

The broader `tests/cron/` run passed 983 tests and exposed 15 unrelated existing Windows/platform or timing failures in `test_file_permissions.py`, `test_cleanup_timeout.py`, `test_cron_workdir.py`, `test_media_delivery_parity.py`, `test_script_claim_heartbeat.py`, and `test_monitor_kind.py`. The cleanup timeout failure passed on isolated rerun; the focused changed suites remained green.

### Code

- [x] I've run the repository test entry on the relevant suites and all focused tests pass

- [x] Documentation is N/A; the code comment records the executor context contract
- [x] `cli-config.yaml.example` is N/A; no config keys changed
- [x] `CONTRIBUTING.md` and `AGENTS.md` are N/A; no architecture or workflow changed
- [x] Cross-platform impact considered; this uses Python ContextVars and executor APIs uniformly
- [x] Tool descriptions and schemas are N/A; no tool surface changed

## Screenshots / Logs

N/A; the regression asserts the persistence target directly.

### DIFF
--- a/cron/scheduler.py
+++ b/cron/scheduler.py
@@ -6368,7 +6368,14 @@ def run_job(
 
             if _session_db_timeout > 0:
                 _session_db_pool = concurrent.futures.ThreadPoolExecutor(max_workers=1)
-                _session_db_future = _session_db_pool.submit(SessionDB)
+                # ThreadPoolExecutor does not copy ContextVars. Preserve the
+                # multiplex ticker's per-profile HERMES_HOME override so the
+                # bounded worker opens the job owner's state.db, not the
+                # dashboard/gateway process profile's store (#97489).
+                _session_db_context = contextvars.copy_context()
+                _session_db_future = _session_db_pool.submit(
+                    _session_db_context.run, SessionDB
+                )
                 try:
                     _session_db = _session_db_future.result(timeout=_session_db_timeout)
                 except concurrent.futures.TimeoutError:
--- a/tests/cron/test_sessiondb_init_hang.py
+++ b/tests/cron/test_sessiondb_init_hang.py
@@ -75,6 +75,53 @@ def _session_db_executor(timeouts: list, *, instant_timeout: bool = True):
 
 
 class TestSessionDbInitTimeout:
+    def test_bounded_init_preserves_profile_home_scope(self, tmp_path, monkeypatch):
+        """The timeout worker must open the owning profile's state store."""
+        from hermes_constants import (
+            get_hermes_home,
+            reset_hermes_home_override,
+            set_hermes_home_override,
+        )
+
+        root_home = tmp_path / "dashboard"
+        profile_home = tmp_path / "profiles" / "north-caribbean"
+        root_home.mkdir()
+        profile_home.mkdir(parents=True)
+        monkeypatch.setenv("HERMES_HOME", str(root_home))
+        monkeypatch.setenv("HERMES_CRON_SESSION_DB_TIMEOUT", "10")
+        opened_homes = []
+        fake_db = MagicMock()
+
+        def _scoped_session_db():
+            opened_homes.append(get_hermes_home())
+            return fake_db
+
+        job = {"id": "profile-scoped-db", "name": "test", "prompt": "hello"}
+        with patch("cron.scheduler._hermes_home", profile_home), \
+             patch("cron.scheduler._resolve_origin", return_value=None), \
+             patch("hermes_cli.env_loader.load_hermes_dotenv"), \
+             patch("hermes_cli.env_loader.reset_secret_source_cache"), \
+             patch("hermes_state.SessionDB", side_effect=_scoped_session_db), \
+             patch(
+                 "hermes_cli.runtime_provider.resolve_runtime_provider",
+                 return_value=_RUNTIME,
+             ), \
+             patch("run_agent.AIAgent") as mock_agent_cls:
+            mock_agent = MagicMock()
+            mock_agent.run_conversation.return_value = {"final_response": "ok"}
+            mock_agent_cls.return_value = mock_agent
+
+            token = set_hermes_home_override(profile_home)
+            try:
+                success, _output, final_response, _error = run_job(job)
+            finally:
+                reset_hermes_home_override(token)
+
+        assert success is True
+        assert final_response == "ok"
+        assert opened_homes == [profile_home]
+        assert mock_agent_cls.call_args.kwargs["session_db"] is fake_db
+
     def test_run_job_does_not_hang_when_sessiondb_init_wedges(self, tmp_path, monkeypatch):
         """run_job proceeds without a session store when SessionDB init times out."""
         monkeypatch.setenv("HERMES_CRON_SESSION_DB_TIMEOUT", "0.2")


andrexibiza (commented):
Thanks for isolating the nested executor boundary and adding a focused two-home regression. I reviewed exact head `4d38b1c2cc38910a0dfe9346d9cd920c8166d4c2` against current `main@ac6c8028e00d01ee2f299ba7fd03329c7f10382d`, the actual diff, #97489, the current logging path, the prior profile-scope work in #87645/#87575, adjacent scheduler/delivery ownership work, existing review history, and exact-head CI. The `copy_context()` / `ctx.run(SessionDB)` mechanism itself is sound for the bounded SessionDB worker, but I see two blockers before this can truthfully close the reported class.

### P1 — #97489 is only half fixed: the process-bound log sink still targets the ticker profile

The production report in #97489 is explicitly a **two-sink** isolation failure: the wrong profile receives both the cron session in `state.db` **and** the run's entries in `logs/agent.log`. This diff repairs only the first sink.

Current `hermes_logging.setup_logging()` resolves `get_hermes_home()` when logging is initialized and installs a concrete `FileHandler(log_dir / "agent.log")`. There is no per-record ContextVar lookup that would make an already-installed dashboard/gateway handler follow the owner profile later. So after this patch the nested SessionDB can correctly open profile A while the same multiplex execution still appends its log records to process profile B.

That leaves the advertised profile-persistence invariant split across two owners and makes `Closes #97489` stronger than the implementation. Please either:

1. carry exact owner-profile identity through the persistent logging sink as well, with a two-home regression proving a unique cron marker appears in the owner's `agent.log` and not the ticker/dashboard log; or
2. narrow the closure relationship so #97489 remains open (or is split with an explicit remaining issue) until the log side is repaired.

Given the current closing claim and the issue's evidence, I would strongly prefer one end-to-end invariant: **a multiplex cron execution may persist session state and durable logs only under the owning profile**.

### P1 — reconcile the prior owner of this exact SessionDB correction before landing

The production hunk here is not novel: open #87645 by Christopher-Schulze already contains the same `contextvars.copy_context()` → `_ctx.run(SessionDB)` correction for #87575, alongside broader profile-runtime scoping work. #97504 is **not** a wholesale duplicate—the fresh #97489 incident and this focused nested-executor regression are useful new evidence—but the shared implementation needs an explicit supersession/salvage decision so we do not land two independent owners for the same boundary or erase the earlier contributor lineage.

If #97504 is the narrower landing vehicle, please preserve/credit Christopher-Schulze's prior SessionDB slice and keep #87645's broader runtime-scope work distinct. If #87645 lands first, rebase/drop the duplicated production hunk here and retain the unique #97489 witness plus the still-missing logging repair. Either shape is fine; silently merging both is not.

The nearby work is complementary, not duplicate: #91953 addresses which scheduler/ticker is allowed to execute when a live gateway owns the profile, while #95899 addresses owning-profile adapters/credentials for multiplex delivery. Those are the other sides of the same qualified-identity boundary, but neither substitutes for exact persistence ownership here.

### Exact-head acceptance is still red

At reviewed head `4d38b1c2...`, [CI run 33219720909](https://github.com/NousResearch/hermes-agent/actions/runs/33219720909) is failed: the Python `Run tests` job and aggregate `All required checks pass` job are red. The contributor check, lint, platform-specific checks, and E2E jobs are green; [Docker 33219720043](https://github.com/NousResearch/hermes-agent/actions/runs/33219720043) and [Nix 33219720116](https://github.com/NousResearch/hermes-agent/actions/runs/33219720116) are green. This commit does not yet have an all-green hosted acceptance receipt.

What does check out: the nested executor really is a ContextVar-loss boundary, the chosen propagation mechanism is the right primitive for SessionDB construction, and the new regression is non-vacuous because it gives process and owner scopes different homes and asserts the actual constructed handle follows the owner. Complete the logging side, reconcile the prior implementation owner, and get the exact commit fully green; then this becomes a clean end-to-end profile-persistence repair rather than a partial sink fix. 🚀

fangliquanflq:
Thanks for the detailed review. I confirmed each point against the current issue, source, overlapping PRs, and exact-head CI:

- **Logging ownership is a valid gap.** #97489 reports both `state.db` and `logs/agent.log` crossing the profile boundary, while this PR only changes the bounded `SessionDB` initialization path and its regression test. The existing file handlers are bound to the home selected by `setup_logging()`, so the current diff cannot truthfully close the full two-sink report.
- **The prior implementation ownership also needs coordination.** #87645 by Christopher-Schulze predates this PR and contains the same `copy_context()` / `ctx.run(SessionDB)` production correction as part of broader profile-runtime scoping work. Since #97570 has now appeared with both the SessionDB correction and profile-aware queued-log routing, I am not adding another independent logging implementation or inventing attribution metadata here. A maintainer needs to select the landing vehicle and preserve the relevant contributor lineage.
- **The exact-head CI failure is real but unrelated to this diff.** Run 33219720909 completed with 39,790 tests passing and one failure: `tests/test_managed_runtime_resolution.py::test_allowlist_has_no_stale_entries` raised `FileNotFoundError` while `Path.rglob()` traversed `plugins/web/brave_free/__pycache__`. This PR changes only `cron/scheduler.py` and `tests/cron/test_sessiondb_init_hang.py`; it does not touch that scanner, plugin, or pycache lifecycle. The later #97570 run passed the Python test lane and all required CI checks.

I have therefore not pushed code or changed the closing relationship in this pass: doing either before the overlapping implementation/credit decision would risk creating a third owner for the same boundary or discarding valid prior work.

PR #96748, opened 2026-08-28
fix(gateway): drop stale model/provider keys in session model_config (resume routed to wrong provider)

ahrazzle:
## What

`_runtime_model_config` (TUI/desktop gateway persist path) merges the agent's current identity onto the row's existing `model_config` JSON. For `model` and `provider` it only SET the key when the agent attribute was truthy — while `base_url` / `api_mode` / `reasoning_config` / `service_tier` already **deleted** stale values when falsy. When an agent rebuilt with an empty provider (inheriting the profile default) was persisted, the previous provider/endpoint survived in `model_config` while `_persist_live_session_runtime` updated the `model` column separately.

On resume, `_stored_session_runtime_overrides` reads the model from the column but the provider/endpoint from `model_config` — so the chat silently routed to the stale provider (e.g. a VeniceAI/empero endpoint) under a model that should run on the profile default.

This is the same stale-key bug class the CLI path already fixed with or-None deletion in `_persist_model_switch_to_session` (see `tests/cli/test_resume_model_restore.py::test_persist_model_switch_clears_stale_route_keys`); the gateway writer never got the symmetric treatment. The fix applies delete-on-falsy to `model` and `provider` too, matching the existing handling of the other identity keys.

## Evidence (real incident)

During a bulk provider rotation across 8 profiles / 174 chat sessions, the audit found **4 live rows** where `sessions.model` was already `deepseek/deepseek-v4-flash-0731` but `model_config` still carried `provider: stealth-ox-alpha` (VeniceAI) / `empero` with `base_url` + `api_mode`. Config.yaml was correct; the desync was purely in the session rows, and would have routed those resumes to the wrong endpoint.

## Reproduction

```python
from tui_gateway.server import _runtime_model_config
from types import SimpleNamespace

agent = SimpleNamespace(model="deepseek/deepseek-v4-flash-0731", provider="",
                        base_url="", api_mode="", reasoning_config=None,
                        service_tier=None)
existing = {"model": "deepseek/deepseek-v4-flash-0731",
            "provider": "stealth-ox-alpha",      # stale
            "base_url": "https://api.venice.ai/api/v1", "api_mode": "chat_completions"}
config = _runtime_model_config(agent, existing)
# before: config["provider"] == "stealth-ox-alpha"  (stale survives)
# after:  "provider" not in config                  (deleted — resume falls back)
```

## Tests

New `TestRuntimeModelConfigDropsStaleKeys` in `tests/tui_gateway/test_custom_provider_session_persistence.py`:
[...]

Issue #96745, opened 2026-08-28
feat: session model audit / bulk reset CLI (sessions pinned to outdated providers)


ahrazzle:
## Feature Description

Add a supported way to audit and bulk-reset the per-session model/provider overrides stored in each profile's `state.db` `sessions` table. After a provider rotation or a global config model change, every existing chat that ever made a session-scoped model pick stays pinned to the old model/provider on resume — and the only way to fix them today is raw SQL against the DB.

Proposed surface:

- `hermes sessions list --model <glob>` / `--provider <name>` — audit: show sessions whose stored model/provider deviate from the profile default (or match a filter), with their source (TUI/desktop/CLI/gateway), profile, last activity.
- `hermes sessions reset-model [--to <model>] [--provider <name>] [--dry-run] [--all]` — bulk reset of session-scoped model overrides back to a target (default: the profile's configured default). Must:
  - back up `state.db` before writing (cp or sqlite backup API),
  - support `--dry-run` printing the exact rows that would change,
  - only touch `sessions` rows, never `cron_*` run records,
  - write `model`, `billing_provider`, and `model_config` consistently (no column/JSON desync),
  - print a read-back verification (count of remaining non-target sessions per profile).

## Motivation

Real incident: a global provider switch across 8 profiles / 174 chat sessions left a mix of pinned models (`meituan/longcat-2.0:free`, `stealth-ox-alpha/custom`, `Qwen/Qwen3.8-Flash-Next-FP8` via empero, `stepfun/step-3.7-flash:free`, `z-ai/glm-5.2`, `hy3-free`, …). Config.yaml was correct; the sessions were correct per their own last-used state. Restoring uniformity required a hand-written SQL script (backed up first) against all 8 `state.db` files — exactly what a supported CLI should do safely, with a dry run, a backup, and a read-back verification.

That incident also surfaced the underlying desync bug (model column vs `model_config` provider) — fixed separately in #96748.

## Proposed Solution

Two `hermes sessions` subcommands (audit + reset) as above. The session-scoped override mechanism is deliberate (a resumed chat should restore what it used), so this is not about removing that behavior — it's about giving users a safe tool to *change* it in bulk when the environment rotates.

## Alternatives Considered

- **Manual SQL / `sqlite3`** — what was done this time; unsafe without backup discipline, error-prone (desync risk), no verification.
[...]

PR #97510, opened 2026-08-28
feat(sessions): session model audit/reset CLI (--model/--provider filters + bulk re-pin)

ahrazzle:
"## What\n\nCLI visibility and control for per-chat model state. Sessions-only, per Teknium1's direction on the closed PR \u2014 the cron half (cron list model pins) built on the fail-closed drift guard removed in #106499, so it is dropped here.\n\n**1. `sessions list --model/--provider` \u2014 audit filters over the EFFECTIVE identity**\n\nThe `model_config` JSON wins on resume (`_stored_session_runtime_overrides` reads provider/endpoint from it), so the filter reads the JSON first and falls back to the `model` column / `billing_provider`. That is what makes `sessions list --provider nous` surface the desync class \u2014 a chat whose model column looks current but whose JSON still pins an old provider/endpoint.\n\n**2. `sessions repair --model TARGET [--provider TARGET]` \u2014 bulk re-pin**\n\nRewrites every non-cron session whose effective model/provider deviates from the target. Safety by construction:\n- timestamped `state.db` backup first (`--no-backup` opts out),\n- `cron_%` run records never touched (they are history, not chats),\n- lineage/tuning keys in `model_config` preserved verbatim (`_branched_from`, `_delegate_from`, `reasoning_config`, `service_tier`, `max_iterations`, `max_tokens`, `gateway_runtime`),\n- stale endpoint keys (`base_url`/`api_mode`) dropped for built-in provider targets,\n- `--dry-run` lists the plan without writing; `--all` also pins NULL-model sessions that inherit the profile default,\n- auxiliary models remain config.yaml-only (never session rows).\n\n## Evidence (real incident)\n\n8-profile Hermes deployment, 174 chat sessions. After a provider rotation, `config.yaml` was correct on all 8 profiles yet some chats still routed to the old endpoint on resume. Root cause: per-session overrides in `state.db` \u2014 the `model` column said the new model while the `model_config` JSON still pinned the old provider, or the column itself was stale:\n\n```\n20260828_055654_642a85|deepseek/deepseek-v4-flash-0731|nous|{\"model\": \"glm-5.3-flash\", \"provider\": \"empero\", \"base_url\": \"https://free.empero.org/v1\", ...}\n```\n\nThe only way to audit or fix these was raw SQL on `state.db` \u2014 no CLI existed.\n\n## Reproduction\n\n```bash\n# Audit: which chats deviate from the profile default?\nhermes sessions list --provider nous --limit 50        # desync rows show up here\n\n# Bulk re-pin after a provider switch (dry-run first, then real)\nhermes sessions repair --model deepseek/deepseek-v4-flash-0731 --provider nous --dry-run\nhermes sessions repair --model deepseek/deepseek-v4-flash-0731 --provider nous\n```\n\n## Tests\n\n`tests/hermes_cli/test_sessions_model_reset.py` (8 tests, real-DB E2E against a temp `HERMES_HOME`):\n\n- list `--model` filter matches the JSON-pinned model, not just the column;\n- list `--provider` filter catches the desync class (correct column, stale JSON provider);\n- repair rewrites deviating rows, preserves `_branched_from`/`reasoning_config`, drops stale `base_url`/`api_mode`, leaves cron records and already-correct rows untouched;\n- repair `--dry-run` writes nothing;\n- repair default skips NULL-model and already-matching rows; `--all` pins NULL rows;\n- `--model`-only keeps provider routing untouched;\n- `--no-backup` skips the copy.\n\nSuite results (via `scripts/run_tests.sh`): the 8 new tests pass; adjacent session CLI files (listing, pin, error-exit-codes, delete, filters) 38 passed, 0 failed.\n\n## Existing affected rows\n\nUsers with desynced rows do NOT need manual action for the bug itself \u2014 the writer-side fix (#96748) self-heals rows on the next live persist. This PR is the CLI to find and reset them on demand (e.g. immediately after a rotation, or to force-clear stale endpoint keys with `--all`).\n\nRelated: https://github.com/NousResearch/hermes-agent/issues/96745\n"

### DIFF
--- a/hermes_cli/cron.py
+++ b/hermes_cli/cron.py
@@ -192,6 +192,13 @@ def _job_rows(job: Dict[str, Any]) -> List[tuple[str, str]]:
     monitor_source = job.get("monitor_script") or job.get("monitor_url")
     mon_state = job.get("monitor_state") or {}
     latest_execution = job.get("latest_execution") or {}
+    # Model/provider pin: an unpinned job runs on the profile default and can drift-skip
+    # silently when that default changes. Show the pin so `cron list` is the first place
+    # a user sees it, not a jobs.json read.
+    _model_val = job.get("model")
+    _provider_val = job.get("provider")
+    model_display = _model_val if _model_val else color("(profile default)", Colors.DIM)
+    provider_display = _provider_val if _provider_val else color("(profile default)", Colors.DIM)
     optional = [
         ("Skills", ", ".join(skills) if skills else ""),
         ("Script", job.get("script")),
@@ -211,6 +218,8 @@ def _job_rows(job: Dict[str, Any]) -> List[tuple[str, str]]:
         ("Schedule", job.get("schedule_display", job.get("schedule", {}).get("value", "?"))),
         ("Repeat", f"{repeat_info.get('completed', 0)}/{repeat_times}" if repeat_times else "∞"),
         ("Next run", job.get("next_run_at", "?")),
+        ("Model", model_display),
+        ("Provider", provider_display),
         ("Deliver", deliver if isinstance(deliver, str) else ", ".join(deliver)),
     ] + [(label, value) for label, value in optional if value]
 
--- a/hermes_cli/sessions_cmd.py
+++ b/hermes_cli/sessions_cmd.py
@@ -43,6 +43,49 @@ def _confirm_prompt(prompt: str) -> bool:
         return False
 
 
+#: Operational keys that must survive a model-pin rewrite. Identity keys
+#: (model/provider/base_url/api_mode) are rewritten or dropped; everything
+#: else in a row's model_config JSON is lineage or tuning and is preserved
+#: verbatim (dropping _branched_from/_delegate_from severs branch ancestry).
+_MODEL_CONFIG_PRESERVE_KEYS = frozenset(
+    {
+        "reasoning_config",
+        "service_tier",
+        "max_iterations",
+        "max_tokens",
+        "gateway_runtime",
+        "_delegate_from",
+        "_branched_from",
+    }
+)
+
+
+def _parse_model_config(raw):
+    """Parse a sessions.model_config JSON string; {} on absent/garbage."""
+    if not raw:
+        return {}
+    try:
+        parsed = json.loads(raw) if isinstance(raw, str) else raw
+        return parsed if isinstance(parsed, dict) else {}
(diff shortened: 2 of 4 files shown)


ahrazzle:
Operating evidence for a config-layer sibling gap this PR does not cover.

Real incident (2026-08-30, six-agent team): a chat began answering in Chinese,
with the underlying symptom "slow / unresponsive primary." Session audit came
back clean — every session row was already nous/deepseek/deepseek-v4-flash-0731.
The actual vector was config.yaml `fallback_providers`, still pinned to
`opencode-free/muse-spark-1.2-contributor-free` for that profile. When the
nous/deepseek primary was slow or timed out, the gateway fell back to that
chain entry and it answered in Chinese. Aligning `fallback_providers` back to
nous/deepseek fixed it.

Consequence for this PR: after a model/provider switch, a user who runs
`sessions repair --model` sees "0 sessions changed" (rows already on target) and
concludes the migration is complete — while `fallback_providers` still routes
every primary failure to the old provider family. This PR fixes session rows and
cron pins; it does not touch config.yaml fallback. The tool reports "clean" for
a drift vector it never inspects.

Concrete follow-up: a `--check-config` flag on `sessions repair --model`, or a
`doctor` check, that reads the profile's effective `fallback_providers` and
`fallback_model` from config.yaml and reports any chain entry not resolving to
the target provider — reporting only, not rewriting (a fallback chain is a
deliberate choice). `hermes fallback list` shows the chain but does not judge it
against a target.

teknium1:
The fail-closed cron drift guard this PR builds on was removed on main in https://github.com/NousResearch/hermes-agent/pull/106499 (7e4d02fef5b): an unpinned job now runs on its creation snapshot (`model_snapshot`/`provider_snapshot`) whenever the global model/provider differs, so a global model change never skips a cron run and `[drift_skip]`, `drift_alerted` and `cron.model_drift_guard` no longer exist. The cron half (list model pins) is superseded by the guard removal; the sessions model audit/reset half is unrelated to cron and would land faster as its own focused PR. Thanks @ahrazzle — feel free to reopen scoped to sessions only.

ahrazzle:
Scoped to sessions only per Teknium1's direction (https://github.com/NousResearch/hermes-agent/pull/97510#issuecomment-5601152461) — the cron list-pin half is dropped (superseded by #106499). The sessions-only branch now lives at #106650.

PR #97512, opened 2026-08-28
fix(compression): enforce total ceiling across attempts

### LINKED ISSUE #97488: [Bug]: Lean compaction ceiling timeout leaves detached workers and can falsely auto-reset sessions
## Bug description

On a large gateway session, a healthy lean compaction can exceed the host-owned 600-second total ceiling even though every provider request succeeds. Hermes then returns without compression, but the detached worker continues issuing sequential digest requests and a later compression attempt can start against the unchanged session.

A later incident on the same failure path confirmed a second, more destructive bug: after the host times out and records a compression cooldown, a real upstream `context_length_exceeded` can trigger reactive compression while that cooldown is still active. `_compress_context()` returns the original list because automatic compression is temporarily blocked, but the overflow handler treats that no-op as permanent `compression_exhausted` and the gateway auto-resets the session.

This was found while correlating Hermes with the OpenAI-compatible routes exposed by [Codex Pooler](https://github.com/icoretech/codex-pooler). No private transcript content, credentials, endpoint hostnames, account identifiers, session identifiers, or request identifiers are included below.

## Environment

- Hermes Agent v0.20.6 on Linux with Python 3.11
- Observed installed commit `420e156bf36e61d047fa526b9e0fcf30b7f26f90`
- Gateway session hygiene followed by the general Hermes compression path
- OpenAI-compatible `openai-api` provider through Codex Pooler
- `model.context_length: 828400`, explicitly configured by the operator
- `compression.tail_mode` unset, therefore using the current default `lean`
- Effective host limits:
  - hygiene inactivity window: 30 seconds
  - hygiene total ceiling: 600 seconds
  - normal context-compression inactivity window: 120 seconds
  - normal context-compression total ceiling: 600 seconds
- `auxiliary.compression.timeout: 900`

The auxiliary timeout is larger than the host's adoption window and cannot make the complete multi-call compaction commit successfully.

The unconfigured tail mode is relevant because [`6e5413844e`](https://github.com/NousResearch/hermes-agent/commit/6e5413844e77ffdf0b8a300b946bdbac934a72f6) changed the default from `legacy` to `lean`. Lean mode adds identifier-preserving chunk digests after the main summary. The default flip is a strong temporal candidate for exposing the issue, although a controlled `legacy` versus `lean` reproduction would still be needed to prove that part causally.

## Reproduction shape

1. Grow a tool-heavy gateway session until session hygiene or normal preflight triggers compression.
2. Use lean compaction with enough serialized history to require many identifier-preserving chunk digests.
3. Let the provider successfully complete the main summary and every digest request, but with aggregate latency greater than 600 seconds.
4. Observe the host hit its total ceiling and continue without compression.
5. Observe the original worker continue making digest calls after the host returned.
6. Trigger another automatic compression attempt against the unchanged session.
7. Observe overlapping workers followed by `commit_fence_cancelled` when older candidates eventually reach the commit boundary.
8. Let a main model request exceed the real upstream context limit while the just-recorded compression cooldown remains active.
9. Observe the overflow recovery call `_compress_context()`, receive the original list because automatic compression is cooldown-blocked, classify the no-op as `compression_exhausted`, and auto-reset the session.

### Phase 1: healthy provider calls exceed the compaction ceiling

One affected logical compaction performed:

- one main summary request
- 23 sequential lean chunk-digest requests from `ContextCompressor._build_chunk_digests()`
- 24 successful provider requests in total

The main summary completed after approximately 163 seconds with visible text output and a completed terminal event. The host nevertheless returned after approximately 615 seconds because the 600-second total ceiling had been exceeded.

The user-facing warning was:

```text
Context compression timed out after 30.0s with no output from the summary model
```

That message did not describe the observed exit condition:

- the actual trigger was the 600-second total ceiling, not 30 seconds of inactivity
- the main summary had already completed with output
- progress events continued throughout the wait

The detached worker continued for approximately another 12 minutes, completed the remaining digest work, and was rejected at commit with:

```text
commit_status="aborted"
failure_class="commit_fence_cancelled"
```

A second compression attempt started before that worker finished. Subsequent attempts repeated the same pattern while the durable transcript remained unchanged and the committed compression count remained zero.

### Phase 2: real context overflow becomes false permanent exhaustion

After another 600-second preflight compression timeout, Hermes sent the unchanged oversized turn. The matching Codex Pooler request had this metadata-only shape:

- public `/v1/responses`, downstream and upstream HTTP SSE
- no Codex continuity session and no `previous_response_id`
- 3,540,973 JSON bytes on both the client and upstream payloads
- 1,904 input entries on both sides:
  - 242 messages
  - 831 function calls
  - 831 function-call outputs
- initial upstream HTTP 200 followed by terminal `context_length_exceeded`
- zero text deltas, one attempt, zero retries
- final provider usage unknown; JSON-shape/reservation estimate approximately 885K input tokens

Codex Pooler did not add bytes or entries and relayed the upstream terminal failure.

Hermes then attempted overflow recovery, but the previous compression timeout had installed a cooldown. The automatic guard returned the original message list without running another compression. The overflow handler accepted progress only if message count decreased, estimated tokens fell by at least 5%, or a smaller context limit was learned. None happened, so it emitted:

```text
Context length exceeded (1,909,089 tokens). Cannot compress further.
```

The displayed `1,909,089` is Hermes' local post-no-op `estimate_messages_tokens_rough(messages)`, not provider usage and not the model's context maximum.

The result returned `compression_exhausted=true`, and the gateway immediately created a fresh session and emitted:

```text
Session auto-reset — the conversation exceeded the maximum context size and could not be compressed further
```

This was a false permanent-exhaustion decision. Compression was temporarily blocked by cooldown, not proven impossible.

## Codex Pooler boundary

The incident contains a real upstream context-limit error, but the observed Pooler boundary does not show payload inflation or hidden history reconstruction:

- client and upstream JSON sizes and item counts were identical
- the failing request was session-less HTTP SSE
- no `previous_response_id`, Codex session, or upstream websocket bridge was involved
- unrelated summary and digest requests completed normally in the same time window
- both Pooler application replicas remained healthy with zero restarts

The recent native websocket/continuation compaction work therefore did not execute on the failing request shape. Codex Pooler is part of the error path because it relayed the real terminal `context_length_exceeded`; the destructive reset decision happened afterward inside Hermes.

## Expected behavior

- A total-ceiling timeout should report the actual total-ceiling trigger and elapsed duration.
- A warning must not claim "no output" when provider output was observed.
- Once the host cancels a pre-commit candidate, that worker must stop issuing new main-summary or digest requests.
- The same unchanged session must not start another automatic compression while an older worker is still active.
- The wall-clock budget should cover the complete compaction plan, including the main summary and all lean digests.
- The original transcript must remain authoritative until a candidate commits successfully.
- If overflow recovery is blocked by a live compression cooldown, Hermes must return a deferred/retryable result, not `compression_exhausted`.
- A temporary cooldown/no-op must never auto-reset the session.

## Suspected root causes

There are several interacting gaps:

1. `gateway/run.py` formats the hygiene warning from `hygiene_timeout_seconds` even when the exit condition is `hygiene_total_ceiling_seconds`.
2. The commit fence prevents a late durable mutation, but cancelling the fence does not stop the already-running executor worker.
3. `ContextCompressor._build_chunk_digests()` performs independent digest calls sequentially and does not check a shared compression cancellation/deadline between calls.
4. Releasing the durable compression lock before the detached worker exits allows a new attempt against the same unchanged session.
5. `auxiliary.compression.timeout` is request-scoped, while the outer 600-second ceiling is pipeline-scoped; a single compaction can contain many successful requests.
6. `agent/conversation_compression.py` can return the original list when an automatic compression guard such as cooldown is active.
7. `agent/conversation_loop.py` distinguishes lock contention through `compression_deferred`, but does not preserve the same temporary/deferred meaning for cooldown-blocked overflow recovery.
8. The unchanged list is therefore interpreted as permanent exhaustion and propagated as `compression_exhausted=true` to `gateway/run.py`, which resets the session unconditionally.

### Total-ceiling and detached-worker coverage

1. Use one main summary plus many healthy digest calls whose aggregate latency exceeds the outer deadline.
2. All provider calls return successful content; no keepalive-only or stalled transport is required.
3. The host reaches the total ceiling during the digest map phase.
4. No new digest request starts after cancellation wins.
5. A second automatic attempt cannot acquire the session while the old worker is active, or a durable source/strategy brake prevents immediate re-entry.
6. The transcript cannot be mutated by the late candidate.
7. The warning distinguishes inactivity timeout from total-ceiling exhaustion and reports whether semantic output was observed.
8. A shared deadline is checked before and after each digest call.

### Context-overflow and cooldown coverage

1. Let preflight compression hit its total ceiling and install a cooldown.
2. Make the following main request return a real structured `context_length_exceeded`.
3. Assert that overflow recovery recognizes the active cooldown as temporary deferral.
4. Assert that `compression_exhausted` remains false and the existing session is not reset.
5. Assert that the failed oversized turn is not persisted and does not grow the session.
6. Assert that a later attempt can retry after cooldown expiry, while manual forced compression can bypass the cooldown intentionally.
7. Keep the existing guarantee that lock-contention deferral also preserves the session.

Bounded digest concurrency proposed in [#97180](https://github.com/NousResearch/hermes-agent/issues/97180) would reduce wall time, but cancellation, retry exclusion, and deferred-state propagation are still required for correctness.

## Related issues

- [#96707: content-free auxiliary events defeat compression inactivity timeout](https://github.com/NousResearch/hermes-agent/issues/96707) covers false semantic progress. This reproduction includes successful content-bearing responses and instead exhausts the full pipeline ceiling.
- [#96775: interrupted stalled compression re-enters the same strategy](https://github.com/NousResearch/hermes-agent/issues/96775) covers durable backoff after interruption. This reproduction includes the hard total-ceiling path and a cooldown-blocked overflow recovery.
- [#96953: compression commit-fence cancellation and session-hygiene retry livelock](https://github.com/NousResearch/hermes-agent/issues/96953) covers restart/stop invalidation. This reproduction requires no user cancellation or restart.
- [#97180: bounded concurrency for lean chunk digests](https://github.com/NousResearch/hermes-agent/issues/97180) addresses the sequential latency amplifier but not detached-worker cancellation or false exhaustion by itself.

fangliquanflq:
## What does this PR do?

Stops a compression attempt from outliving its host-owned total ceiling and prevents a superseding attempt from overlapping the same unchanged session while the timed-out worker is still finishing its single auxiliary request.

### Symptom

Gateway session hygiene or the in-agent compression wrapper can return after the configured total ceiling while the compression worker is still active. Without a shared deadline and retained lease, another automatic attempt can start against the unchanged session. Timeout messages can also misreport a healthy-but-slow request as producing no output.

### Impact

Large sessions can consume extra provider time in overlapping attempts, remain uncompressed after a late result loses at the commit fence, and receive misleading provider-health guidance.

### Bug Cause

**Trigger:** aggregate compression latency exceeds the host-owned total ceiling.

**Causal chain:**
1. The host wait reaches the total ceiling while the auxiliary request is still active or while summary progress is being observed.
2. The commit fence prevents a late durable mutation, but the worker did not share the host's wall-clock deadline and the total-ceiling path could release its durable lease before worker cleanup.
3. A new attempt can then acquire the unchanged session while the old attempt is still alive, and the timeout is reported as inactivity even when progress occurred.

**Why it is wrong:** commit fencing protects the mutation boundary but does not by itself coordinate attempt lifetime or classify why the host stopped waiting.

**Working sibling / contrast:** inactivity timeouts still release the holder-qualified lease promptly because a provider call may never return. Total-ceiling cancellation retains the lease until worker cleanup because the worker observes the shared deadline around its single auxiliary request.

**Ruled out:** provider transport failure is not required. Successful content-bearing responses can still arrive after the shared total deadline.

### Fix

- Add a monotonic total deadline and progress-observed state to `CompressionCommitFence`.
- Check the shared cancellation signal before and after the single auxiliary summary request, and reject commit admission after the deadline.
- Retain a total-ceiling candidate's durable session lease until worker cleanup so retries cannot overlap the unchanged source.
- Bound host wait slices by the remaining total budget.
- Distinguish total-ceiling exhaustion from inactivity in gateway and in-agent logs, cooldown reasons, and user warnings.

## Related Issue

Closes #97488

## Type of Change

## Changes Made

- `agent/context_compressor.py` - observe cooperative cancellation before and after the single auxiliary request.
- `agent/conversation_compression.py` - share the total deadline, retain total-timeout leases, and propagate structured timeout cause.
- `gateway/run.py` and `run_agent.py` - enforce remaining total budget and emit cause-accurate timeout warnings.
- `tests/agent/` and `tests/gateway/` - cover pre-dispatch expiry, late-result discard, retry exclusion, one-request compatibility, and warning text.

## How to Test

```bash
HERMES_TEST_FILE_RETRIES=0 scripts/run_tests.sh tests/agent/test_context_compressor.py tests/agent/test_compression_concurrent_fork.py tests/agent/test_compress_context_progress_timeout.py tests/agent/test_compression_review_76354.py tests/agent/test_lean_single_aux_call.py tests/gateway/test_session_hygiene.py -q
```

### Code

- [x] I've run the repository test entry on the relevant suites

- [x] Relevant documentation update: N/A - no user-facing configuration changed
- [x] `cli-config.yaml.example` update: N/A - no config keys changed
- [x] `CONTRIBUTING.md` or `AGENTS.md` update: N/A - no workflow contract changed
- [x] Cross-platform impact considered: monotonic clocks and threading primitives are platform-neutral
- [x] Tool descriptions/schemas update: N/A - no model tool behavior changed


poisdahl:
Suggesting changes after reviewing the exact PR head on current main.\n\nThe new total-deadline signal cancels the compression owner, but it does not necessarily stop the provider callback that is already running. The cancel_check reaches _run_protected_sync_provider_call, whose documented behavior is to raise AuxiliaryExplicitCancellation in the owner and leave callback(kwargs) running in a daemon thread until the provider timeout.\n\nI reproduced this on current main plus this exact PR head with a blocking provider callback and a CompressionCommitFence(total_ceiling_seconds=0.05):\n\nowner returned with AuxiliaryExplicitCancellation: yes\nprovider callback finished when owner returned: no\nprovider callback finished only after explicit release: yes\n\nThat matters to the lease claim. retain_compression_lock_until_worker_done suppresses the host-side release hook, but compress_context then handles AuxiliaryExplicitCancellation and releases its ordinary compression lease when the owner unwinds. At that point the actual provider callback can still be alive, so a later attempt can overlap the upstream request. The between-digest checks correctly prevent another digest from being dispatched by the old owner; they do not terminate the in-flight request.\n\nA regression should hold the provider callback open past the deadline and assert that no retry admission or lease release occurs until provider_done, not merely until the compression owner future completes. The fix needs either cooperative transport/stream closure, or a bounded admission/lease lifetime tied to actual provider callback completion.\n\nReview setup: exact PR head composed cleanly with current main 66666f6e. The focused daemon-callback probe passed and demonstrates the residual lifetime; I did not run the full suite locally.

fangliquanflq:
Thanks for the detailed probe. I traced this through the exact PR head and the production call topology, and I could not reproduce the claimed lease gap.

The key distinction is that the total-deadline fence is not the cancellation source captured by `_run_protected_sync_provider_call`. In production, `compress_context()` installs the hard-interrupt event in the outer `aux_interrupt_protection(cancel_event=...)` scope, and `_generate_summary()` enters a nested protection scope without replacing it. The fence is installed separately as `_compression_cancelled_check` and is read before/after provider calls and between lean digests. Therefore, when only the total ceiling expires, the compression owner remains blocked until the already-running provider callback returns; it does not unwind through `AuxiliaryExplicitCancellation` while that callback is alive.

That ordering keeps the durable lease intact: the host-side release is suppressed for total-ceiling cancellation, and the owner's `_release_lock()` path cannot run until the provider callback has completed. The explicit hard-cancel path intentionally has different semantics and may detach the provider callback, but that is not the total-ceiling path changed by this PR.

I also ran a focused production-topology probe with a blocking protected provider callback and a 50 ms fence deadline: after the fence expired, both the provider callback and compression owner were still alive; the owner completed only after the provider callback was released. A probe that passes `fence.is_cancelled` directly as the protected provider's `cancel_check` demonstrates a real behavior of the helper, but it does not match this PR's production wiring.

No code change is needed for this point. Validation on the unchanged head:

`HERMES_TEST_FILE_RETRIES=0 scripts/run_tests.sh tests/agent/test_context_compressor.py tests/agent/test_compression_concurrent_fork.py tests/agent/test_compress_context_progress_timeout.py tests/gateway/test_session_hygiene.py -q`

Result: 230 passed, 0 failed.

teknium1:
Note for rebase: PR #98628 (merged, 4f22543509) collapsed lean compaction to a single auxiliary request per attempt, which removes the multi-call digest pipeline this PR's scenarios multiply through. The attempt-lifecycle concern (ceiling/supersession across ATTEMPTS, cf. #97488) may still stand — please rebase and re-scope to what remains.

fangliquanflq:
Thanks for the re-scope note. I merged current main (including 4f22543509) into this branch and removed the obsolete multi-call digest implementation and its scenarios. The PR now preserves only the attempt-lifecycle fix that still applies to the single auxiliary request: the host and worker share the total deadline, late results cannot commit, and a total-ceiling attempt retains its holder-qualified lease until worker cleanup so a superseding attempt cannot overlap the unchanged session. I also updated both gateway and in-agent timeout reporting so progress through the total ceiling is not mislabeled as no output.

The earlier provider-callback concern remains withdrawn: total-ceiling checks surround the provider request but are not wired as the protected transport's hard-cancel signal, so the owner and lease remain alive until that in-flight request returns.

Validation: `HERMES_TEST_FILE_RETRIES=0 scripts/run_tests.sh tests/agent/test_context_compressor.py tests/agent/test_compression_concurrent_fork.py tests/agent/test_compress_context_progress_timeout.py tests/agent/test_compression_review_76354.py tests/agent/test_lean_single_aux_call.py tests/gateway/test_session_hygiene.py -q` — 251 passed, 0 failed.

PR #97514, opened 2026-08-28
fix(telegram): degrade unsupported markdown link targets to display text

### LINKED ISSUE #97497: [Bug]: Telegram exposes malformed Markdown for desktop-only session references
## Bug Description

When an agent refers to a `session_search` result in a Telegram conversation, it can reformat Hermes' desktop-only `@session:<profile>/<id>` reference as a Markdown link with a non-URL target. Telegram then exposes the malformed Markdown instead of showing readable prose.

Observed output:

```markdown
Already underway in [Long-Term Fix Research](Long-Term Fix Research).
```

The user sees the bracket-and-parenthesis syntax in the Telegram message. This is especially visible with Telegram Rich Messages enabled because the rich path receives the raw agent Markdown before the legacy MarkdownV2 formatter.

The problem crosses a surface contract boundary:

- `session_search` currently says its `link` value "renders as a titled link" without qualifying that this behavior was implemented for Hermes Desktop.
- Telegram does not resolve Hermes' internal `@session:` references.
- Telegram's legacy formatter preserves schemeless Markdown destinations such as `(Long-Term Fix Research)`.
- Rich send, final-edit, and draft paths can receive the raw malformed Markdown before that formatter.

## Steps to Reproduce

1. Use Hermes through the Telegram gateway, with Rich Messages enabled.
2. Ask the agent to find or continue work from another session so it calls `session_search`.
3. Have the agent refer to the returned session in its final response.
4. If the model reformats the session title/reference as `[Title](Title)` or `[Title](@session:...)`, inspect the delivered Telegram message.

A deterministic formatter-level reproduction on current code is:

```python
from plugins.platforms.telegram.adapter import TelegramAdapter

adapter = object.__new__(TelegramAdapter)
print(adapter.format_message("Already underway in [Example Session](Example Session)."))
```

The malformed schemeless destination is preserved:

```markdown
Already underway in [Example Session](Example Session)\.
```

Model emission is nondeterministic, but this failure mode has also been observed in the live A/B benchmark reported on #95570: GPT-5.6 Luna emitted `[title](@session:...)` despite the `link_hint`.

## Expected Behavior

On Telegram, an internal or invalid session-link construction should degrade to readable display text:

```text
Already underway in Long-Term Fix Research.
```

Valid supported links such as `[Docs](https://example.com)` should remain clickable. Hermes Desktop should continue rendering valid `@session:` references as clickable titled session links.

## Actual Behavior

Telegram can display malformed Markdown such as:

```markdown
[Long-Term Fix Research](Long-Term Fix Research)
```

The user sees implementation syntax instead of a usable link or clean session title.

## Affected Component

- Gateway
- Tools (`session_search` response contract)
- Bundled Telegram plugin

## Messaging Platform

Telegram

## Debug Report

Not included. The failure is output-level and is grounded below in current-main source, the exact delivered payload, and an existing live benchmark. A full debug share would publish unrelated private configuration and logs.

## Environment

- Hermes Agent v0.20.6 (`upstream 27f29cfc`)
- macOS 26.5.2
- Python 3.11.15
- Telegram gateway with `rich_messages: always`

## Current-source evidence

At current `main` (`3340bbbdad8368e7f3d9f6827d61692adbbd87d6`):

- [`session_search` says the returned value "renders as a titled link" without a platform qualifier](https://github.com/NousResearch/hermes-agent/blob/3340bbbdad8368e7f3d9f6827d61692adbbd87d6/tools/session_search_tool.py#L941-L946).
- [The Telegram MarkdownV2 link conversion preserves any parsed target without validating that it is a supported URL](https://github.com/NousResearch/hermes-agent/blob/3340bbbdad8368e7f3d9f6827d61692adbbd87d6/plugins/platforms/telegram/adapter.py#L9069-L9076).
- [Rich final sends receive the raw content before the legacy formatter](https://github.com/NousResearch/hermes-agent/blob/3340bbbdad8368e7f3d9f6827d61692adbbd87d6/plugins/platforms/telegram/adapter.py#L5785-L5790).
- The same raw-content concern exists for [rich final edits](https://github.com/NousResearch/hermes-agent/blob/3340bbbdad8368e7f3d9f6827d61692adbbd87d6/plugins/platforms/telegram/adapter.py#L6172-L6178) and [rich drafts](https://github.com/NousResearch/hermes-agent/blob/3340bbbdad8368e7f3d9f6827d61692adbbd87d6/plugins/platforms/telegram/adapter.py#L6652-L6657).

## Related Work

- #71162 introduced clickable titled `@session:` references in Hermes Desktop and the `session_search` link contract.
- #95570 moved the link instruction into `link_hint`; [its live benchmark recorded a malformed `[title](@session:...)` emission despite that hint](https://github.com/NousResearch/hermes-agent/pull/95570#issuecomment-5426787177).
- #34448 is adjacent URL-scheme hardening, but its current predicate still permits schemeless targets and its diff targets the retired Telegram adapter path.
- #55073 demonstrates the need to make compatibility decisions consistently across rich send, final-edit, and draft paths.

None of these currently owns the complete observable defect.

## Potential Implementation Direction

This is a proposal, not a required design:

1. Qualify the `session_search` response contract so `@session:` title/click behavior is identified as a Hermes Desktop capability; messaging surfaces should use the returned session title as plain text unless they support an explicit internal-session reference.
2. Add deterministic outbound Markdown-link validation at the shared Telegram boundary before choosing rich or legacy delivery.
3. Degrade schemeless, internal, or unsupported destinations to their display text while preserving supported links.
4. Apply the same behavior to normal sends, finalized edits, and drafts.

## Done When

- Telegram never exposes raw `[label](target)` syntax for schemeless or unsupported session-link targets.
- `[Title](Title)` and `[Title](@session:profile/id)` degrade to `Title` on Telegram.
- Valid supported external links remain clickable.
- Rich send, finalized edit, draft, and legacy MarkdownV2 delivery have regression coverage.
- Hermes Desktop retains clickable titled `@session:` references.

liuhao1024:
## What does this PR do?

Telegram exposes raw `[label](target)` markdown syntax to the user when the target is not a URL Telegram can render. The trigger observed in #97497: models reformat Hermes' desktop-only `@session:<profile>/<id>` references as `[Long-Term Fix Research](Long-Term Fix Research)` or `[Title](@session:profile/id)`. Telegram cannot resolve these destinations, so the delivered message shows the bracket-and-parenthesis syntax instead of readable prose.

This adds a shared outbound check: a markdown link target is only kept clickable when it matches the Bot API `text_link` contract (HTTP(S)/`tg://` URLs). Every other target degrades to its display text. Wired into both delivery paths:

- `format_message` (legacy MarkdownV2): the link-conversion step now emits the escaped display text for unsupported targets instead of preserving the unrenderable destination.
- `_rich_message_payload` (rich path): the raw markdown is scrubbed on the way out. This is the shared payload builder for rich sends, finalized edits, and drafts, so all three rich call sites are covered by the single hook.

Code spans, fenced code blocks, and GFM table blocks hold literal content and are left verbatim — link syntax inside them is content, not a link. Hermes Desktop is untouched: its clickable titled `@session:` rendering lives outside the Telegram adapter.

## Related Issue

Fixes #97497

## Type of Change

## Changes Made

- `plugins/platforms/telegram/adapter.py` — added module-level `_MD_LINK_RE` (the link pattern both paths now share), `_SUPPORTED_LINK_TARGET_RE` (HTTP(S)/`tg://` per the Bot API text_link spec), `_tg_link_target_supported()`, and `_degrade_unsupported_markdown_links()` which protects code/table regions and degrades unsupported links in prose.
- `plugins/platforms/telegram/adapter.py` — `format_message` step 3 (`_convert_link`) degrades unsupported targets to the escaped display text.
- `plugins/platforms/telegram/adapter.py` — `_rich_message_payload` runs `_degrade_unsupported_markdown_links` after newline normalization.
- `tests/gateway/test_telegram_unsupported_link_targets.py` — new regression suite (13 tests).

## How to Test

1. Deterministic formatter-level repro from the issue (previously preserved the schemeless destination):
   ```python
   from plugins.platforms.telegram.adapter import TelegramAdapter
   adapter = object.__new__(TelegramAdapter)
   print(adapter.format_message("Already underway in [Example Session](Example Session)."))
   ```
   Observed result: `Already underway in Example Session\.` — no `](...)` syntax, readable prose only.
2. Rich path: `_rich_message_payload("See [Research](@session:default/abc123).")["markdown"]` returns `See Research.` while `[Docs](https://example.com/x)` and `[settings](tg://settings)` stay clickable links.
3. `uv run --extra dev pytest tests/gateway/test_telegram_unsupported_link_targets.py tests/gateway/test_telegram_format.py tests/gateway/test_telegram_rich_newlines.py tests/gateway/test_telegram_rich_messages.py tests/gateway/test_telegram_send_draft_format.py -q` — Observed result: 97 passed (13 new + 84 existing).

### Code

- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-impact) — or N/A

### DIFF
--- a/plugins/platforms/telegram/adapter.py
+++ b/plugins/platforms/telegram/adapter.py
@@ -503,6 +503,56 @@ def _rich_normalize_linebreaks(text: str) -> str:
     return ''.join(out)
 
 
+# Markdown link syntax shared by both Telegram delivery paths: the legacy
+# MarkdownV2 formatter's link conversion and the rich-message outbound scrub
+# below.
+_MD_LINK_RE = re.compile(r'\[([^\]]+)\]\(([^()]*(?:\([^()]*\)[^()]*)*)\)')
+
+# Bot API link targets are "HTTP or tg://" URLs (MessageEntity text_link
+# spec). Anything else — a schemeless destination like ``[Title](Title)`` or a
+# Hermes-internal ``@session:profile/id`` reference — cannot be rendered as a
+# link by Telegram and is exposed to the user as raw bracket syntax (#97497).
+_SUPPORTED_LINK_TARGET_RE = re.compile(r'(?i)^(?:https?://|tg://)\S+$')
+
+# Regions where link syntax is literal content rather than a link to degrade:
+# inline code spans, plus the fenced-code / pipe-table regions already matched
+# by _RICH_PROTECTED_REGION_RE.
+_LINK_SCRUB_PROTECT_RE = re.compile(
+    r'`[^`\n]+`|' + _RICH_PROTECTED_REGION_RE.pattern,
+    re.MULTILINE,
+)
+
+
+def _tg_link_target_supported(target: str) -> bool:
+    """True if Telegram can render *target* as a clickable link target."""
+    return bool(_SUPPORTED_LINK_TARGET_RE.match(target.strip()))
+
+
+def _degrade_unsupported_markdown_links(text: str) -> str:
+    """Degrade markdown links Telegram cannot render to their display text.
+
+    Models sometimes emit ``[Title](Title)`` or ``[Title](@session:p/id)``
+    when referencing session-search results; Telegram then shows the raw
+    bracket-and-parenthesis syntax instead of readable prose (#97497).
+    HTTP(S)/tg:// targets stay clickable; every other target degrades to the
+    label text. Code spans/blocks and table blocks are left verbatim.
+    """
+    if '[' not in text:
+        return text
+
+    def _degrade(m):
+        return m.group(0) if _tg_link_target_supported(m.group(2)) else m.group(1)
+
+    out: list[str] = []
+    pos = 0
+    for m in _LINK_SCRUB_PROTECT_RE.finditer(text):
+        out.append(_MD_LINK_RE.sub(_degrade, text[pos : m.start()]))
+        out.append(m.group(0))  # protected region kept verbatim
+        pos = m.end()
+    out.append(_MD_LINK_RE.sub(_degrade, text[pos:]))
+    return ''.join(out)
+
+
 # Watchdog bound for `await updater.stop()`. When the underlying TCP socket is
 # in CLOSE-WAIT the PTB polling task is blocked on epoll on the dead socket and
 # never wakes, so an unguarded stop() hangs indefinitely and wedges the whole
@@ -2137,8 +2187,16 @@ class TelegramAdapter(BasePlatformAdapter):
         Single newlines are normalized to Markdown hard breaks so that
         multi-line content (slash-command lists, etc.) renders correctly
         in the rich-message path.  See ``_rich_normalize_linebreaks``.
+
+        Unsupported link targets (schemeless destinations, ``@session:``
+        references) are degraded to their display text on the way out so
+        Telegram never exposes raw ``[label](target)`` syntax (#97497).
         """
-        payload: Dict[str, Any] = {"markdown": _rich_normalize_linebreaks(content)}
+        payload: Dict[str, Any] = {
+            "markdown": _degrade_unsupported_markdown_links(
+                _rich_normalize_linebreaks(content)
+            )
+        }
         if skip_entity_detection:
             payload["skip_entity_detection"] = True
         return payload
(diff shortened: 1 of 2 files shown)


0xble:
I opened #97535 as a complementary producer-side fix for #97497. This PR owns the deterministic Telegram delivery guard; #97535 makes `session_search` stop exposing Desktop-only references on non-Desktop surfaces. The branches merge cleanly, and the combined focused suites pass 141 tests.

liuhao1024:
Thanks for the heads-up and for taking the producer side. The split reads right to me: this PR is the deterministic delivery-side guard (any markdown link target the Telegram Bot API cannot render degrades to display text, regardless of what produced it), while #97535 stops `session_search` from emitting Desktop-only `@session:` references on non-Desktop surfaces in the first place. The two layers are orthogonal, so they can be reviewed and merged independently — defense in depth rather than overlap.

Issue #69283, opened 2026-07-22
test-safety: kanban unit tests silently write to the real HERMES_HOME when hermetic isolation is bypassed


hehehe0803:
## Summary

The kanban board/task unit tests isolate themselves only by setting `HERMES_HOME` to a temp dir via fixtures. If that isolation is ever bypassed, the tests silently write junk boards and junk tasks into the user's real `~/.hermes`. The existing live-system guard does not cover kanban writes, so nothing fails closed. This happened on a developer machine: a test run created 22 junk boards (exact fingerprints of the fixtures in `tests/hermes_cli/test_kanban_boards.py`) and 2 junk tasks (`title="x"`, `assignee="worker"`, from `test_kanban_core_functionality.py`) on a real, active board.

## Mechanism

- The tests rely on `HERMES_HOME=tmp` set by `fresh_home` (`tests/hermes_cli/test_kanban_boards.py`), `kanban_home` (`tests/hermes_cli/test_kanban_core_functionality.py`), and the autouse `_hermetic_environment` (`tests/conftest.py`).
- `get_default_hermes_root()` reads `HERMES_HOME` live on every call (`hermes_constants.py`), so isolation holds ONLY while those fixtures apply.
- If the suite runs without the repo's `tests/conftest.py` (stale second checkout, wrong pytest rootdir, or the test functions invoked directly), `create_board()` / `create_task()` write to the real home, and `connect()` reads the real current-board pointer — so junk tasks land on a real live board.
- The header of `test_kanban_boards.py` already warns about the stale-clone case; the cache-clear line `hermes_constants._cached_default_hermes_root = None` in its fixture is a dead vestige (that attribute never existed).
- The live-system guard in `tests/conftest.py` blocks `os.kill` / `systemctl` / gateway-pid scans only. Kanban DB writes have no guard.

## Proposed fix (fail closed)

Add an autouse kanban write guard to `tests/conftest.py`, as a companion to the live-system guard:

1. After the `HERMES_HOME` redirect, assert `hermes_cli.kanban_db.kanban_home()` resolves under the per-test temp home.
2. Wrap `connect()` / `create_board()` so a target path outside the per-test `HERMES_HOME` raises instead of writing.

This converts a silent write-to-real-home into a loud test error, the same shape as the updater-restart hermeticity fix in #68988. Optionally also delete the dead cache-clear vestige.

I am happy to contribute the guard as a PR if this shape is acceptable.

smfworks:
[...]

PR #97518, opened 2026-08-28
fix(tests): restore sys.modules and sys.path after tests mutate them (15 order-dependent failures)

rodrigogs:
`pytest tests/hermes_cli/ -k kanban` fails 15 tests on a clean tree. Every one of them passes when its own file runs alone. Two globals are behind all 15: a test mutated interpreter-wide state and never changed it back.

## `sys.modules`, and why this is not tidiness

Three kanban fixtures need a fresh `HERMES_HOME` to be read at *import* time, so they dropped `hermes_cli` / `hermes_state` / `hermes_constants` from `sys.modules` and imported again. Dropping the entries and walking away leaves **two copies of the same code** alive for the rest of the session: the one every already-imported test module holds in its globals, and the one `sys.modules` hands to the next importer.

`_kanban_write_guard` — the autouse deny-list from #69283 that keeps the suite off the operator's real `~/.hermes` — patches `connect` on the copy it finds in `sys.modules`. A test module whose global points at the *other* copy therefore calls an **unguarded** `connect`, for the rest of the run, silently. That is precisely how `test_kanban_write_guard.py` was failing: the guard was installed, on the wrong object. The same split let the hook and lifecycle tests observe a module-level registry on one copy while the code under test mutated the other.

`fresh_hermes_modules()` (tests/conftest.py) does the drop, yields, and puts the **originals** back, so the re-import is local to the test that asked for it.

## `sys.path`, and a dependency that stopped existing

`tests/acp/` is the test package for the ACP adapter, and `acp` is also a real third-party dependency. Two files put the **tests** root on `sys.path` at module level — both an off-by-one (`parents[1]` from a file in `tests/hermes_cli/` is `tests/`, not the repo root, which pytest already prepends), both naming the variable `_repo` / `WORKTREE`. From then on `tests/acp/` **shadowed** the dependency and `from acp.schema import ToolKind` raised `ModuleNotFoundError: No module named 'acp.schema'` against a stub with no `schema` submodule.

One of the two files is skipped in its entirety off Windows — collection alone was enough. It is also why `pytest tests/hermes_cli/ tests/acp/` cannot even collect on `main`: 7 collection errors, run interrupted. `cron`, `plugins`, `providers` and `tools` are first-party packages with a same-named directory under `tests/`, so the same insert could have shadowed any of them.

## Verification

| | before | after |
|---|---|---|
| `tests/hermes_cli/ -k kanban` | 15 failed, 294 passed, 1 skipped | **0 failed, 313 passed, 1 skipped** |
| `tests/hermes_cli/ tests/acp/` | 7 collection errors, interrupted | collects |

313 = the same 309 tests, all passing, plus the 4 new ones. The `sys.path` gate is
outside that selection: `tests/test_tests_dir_not_on_sys_path.py` is 2 more, both passing.

Mutation-verified: deleting the restore from `fresh_hermes_modules` brings back 14 of the original failures plus all four new ones — 18 failed.

The `sys.path` gate is not hypothetical either. It **failed on the first run of this branch** and found the second polluter, which is how `test_web_server_host_header.py` came to be in this PR. It ships with a floor test, because a gate guarding a collision that does not exist is theatre: it asserts some directory under `tests/` really does share a name with a repo package.

Not verified: the full `tests/hermes_cli/` suite end to end. On the box available to me it is OOM-killed partway through — a standing memory constraint there, unrelated to these changes (nothing here spawns a process).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,6 +21,7 @@ test runner at ``scripts/run_tests.sh``.
 
 import asyncio
 import atexit
+import contextlib
 import importlib
 import os
 import shutil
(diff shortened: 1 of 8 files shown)


rodrigogs:
Rebased onto current `upstream/main` (GitHub reported CONFLICTING/DIRTY). Single conflict in `tests/conftest.py` was an import-line collision — upstream's `import importlib` and this PR's `import contextlib` are both present now, nothing else touched. Verified on the new base: `test_kanban_module_reimport_isolation.py` + `test_tests_dir_not_on_sys_path.py` 6/6 passed.

rodrigogs:
Rebased onto current `upstream/main` (564aef294). No conflicts.

One adaptation for the new layout: since e3ab65fe8 (`simplify(compat): kanban_db — drop 73 re-exports/aliases`), `connect` lives in `hermes_cli.kanban_db_connect` and `_kanban_write_guard` patches it there; `hermes_cli.kanban_db.connect` is now only a `_PLUGIN_COMPAT_LAZY` shim that emits `HermesPluginCompatWarning` and is slated for removal on 2026-09-14. The two `connect`-based assertions in `test_kanban_module_reimport_isolation.py` now target `kanban_db_connect` directly (same module `test_kanban_write_guard.py` uses), so the test keeps working after the shim goes away. The `sys.modules` identity tests on `hermes_cli.kanban_db` are unchanged.

Verified on the new base, one file per run: `test_kanban_module_reimport_isolation.py` 4/4 (also with `-W error::hermes_cli.plugin_compat.HermesPluginCompatWarning`), `test_tests_dir_not_on_sys_path.py` 2/2, the three modified kanban fixtures' files 2/2 each, `test_web_server_host_header.py` 11/11, `test_kanban_write_guard.py` 3/3. Ruff clean.

Issue #96394, opened 2026-08-27
[Bug]: Desktop makes every approval unanswerable on a single-connection/single-profile install — ambientGatewayOwnsEverySession() is unreachable, so approval.respond always fails owner resolution


Bardioc1977:
# [Bug]: Desktop makes every approval unanswerable on a single-connection/single-profile install — `ambientGatewayOwnsEverySession()` is unreachable, so `approval.respond` always fails owner resolution

## Summary

On a Desktop install with exactly one connection (`local`) and exactly one profile (`default`),
**every approval prompt becomes unanswerable**. Allow/Deny does nothing, the modal cannot be
dismissed, and the turn is stuck permanently:

```
Session owner could not be resolved for "12986933" (approval.respond): no owner route, hint,
connection-tagged row or profile probe named the backend that holds this session, and routing it
to the active gateway would be a guess.
```

Restarting the gateway does not help — the condition is structural, not stateful.

The fail-closed policy itself is right. The defect is its *ambiguity test*: it asks whether a
connection registry **exists**, not whether that registry **names more than one possible owner**.
Since every current Desktop exposes the `connections.list` bridge, the documented escape hatch is
unreachable on all modern installs, including ones where a single local backend is the only thing
that could own the session.

## Root cause

`apps/desktop/src/store/session-owner-resolution.ts:50-52` (v0.20.5, upstream `36b0a96d`):

```ts
export function ambientGatewayOwnsEverySession(): boolean {
  return !hasRegistryTopology() && $profiles.get().length <= 1
}
```

Both conditions must hold. The profile half is satisfied on my install (one profile). The topology
half never is, because of `apps/desktop/src/store/connection-registry-state.ts:9-14`:

```ts
export function hasRegistryTopology(): boolean {
  // The bridge exists before its asynchronous cache load. Treat that window
  // (and a failed list IPC) as registry topology so owner routing fails closed;
  // only an older Desktop without the registry capability is truly legacy.
  return $connectionsRegistry.get() !== null || Boolean(window.hermesDesktop?.connections?.list)
}
```

`window.hermesDesktop.connections.list` is present on any current Desktop, so this returns `true`
regardless of registry contents. `assertSessionOwnerResolved` (`:73-82`) therefore throws whenever
the owner ladder misses — even when the registry proves there is exactly one candidate.

The comment states the intent exactly: *"only an older Desktop without the registry capability is
truly legacy."* That conflates **capability present** with **topology ambiguous**. A registry holding
[...]

PR #97519, opened 2026-08-28
fix(desktop): route approval responses through the runtime event's exact owner

### LINKED ISSUE #97511: fix(desktop): route approval responses through exact runtime event owner
## Summary

In registry topology with multiple profiles, a Desktop approval can render normally but every **Run / Allow / Reject** action fails with `SessionOwnerResolutionError`, even when the inbound `approval.request` event carried an exact `(connectionId, profile)` owner.

This is a sibling of #96394, not the same fallback case. #96394 concerns proving a sole ambient backend. This case has exact owner evidence and should remain fail-closed rather than use ambient routing.

## Root cause

`use-gateway-boot.ts` calls `recordSessionEventScope(event)` before dispatching inbound events. That stores the exact registry scope in `sessionScopeByRuntimeId`. However, `knownOwnerForSession()` never consults that ledger; it only checks persisted tile routes, stored-session owner hints, and hydrated rows.

When an approval is keyed by a runtime ID whose runtime-to-stored binding is absent/stale, `requestForOwnedSession(..., "approval.respond")` therefore throws even though the event source itself proved the owner.

## Reproduction

1. Use Desktop with registry topology and multiple profiles.
2. Receive an `approval.request` on a connection-tagged runtime whose stored-session mapping is unavailable.
3. Click Run, Reject, or an Allow option.
4. Desktop reports `Session owner could not be resolved ... (approval.respond)` and the backend remains blocked.

## Expected

The response routes through the exact owner recorded from the inbound runtime event. Durable stored-session identity must still outrank a colliding stale runtime ID; unknown owners must continue to fail closed.

## Proposed regression coverage

- connection-tagged orphan runtime routes `approval.respond` through its recorded owner;
- a durable stored-session ID collision still uses the stored session owner, not a stale runtime ledger entry;
- an untagged/unknown runtime in multi-profile topology still fails closed.

liuhao1024:
## What does this PR do?

In registry topology with multiple profiles, a Desktop approval can render normally but every **Run / Allow / Reject** action fails with `SessionOwnerResolutionError`, even when the inbound `approval.request` event carried an exact `(connectionId, profile)` owner.

`recordSessionEventScope(event)` (called at event fan-in in `use-gateway-boot.ts`, before dispatch) already stores the exact registry scope per runtime id — but only as an opaque composite string, and `knownOwnerForSession()` never consults it. When an approval is keyed by a runtime id whose runtime-to-stored binding is absent or stale, the three durable rungs (tile route, owner hint, session row) all miss and `requestForOwnedSession(..., "approval.respond")` throws, even though the event source itself proved the owner.

This adds a structured owner twin of that scope ledger, written and cleared with it, and consumes it as the **last** rung of `knownOwnerForSession`:

- A connection-tagged orphan runtime now routes `approval.respond` through the exact owner its events recorded.
- Durable stored identity still outranks it, so a stored-id collision never inherits a stale runtime ledger entry.
- Untagged events (no `connectionId`) record nothing, so unknown owners in multi-profile topology keep failing closed — this is exact-owner evidence, not an ambient fallback (that sibling case is #96394).

## Related Issue

Fixes #97511

## Type of Change

## Changes Made

- `apps/desktop/src/store/session-states.ts` — add `sessionOwnerByRuntimeId` (a `Map<string, SessionOwnerRoute>` twin of the scope ledger), written in `recordSessionEventScope`, cleared in `dropSessionState` / `clearAllSessionStates` alongside the scope ledger, and consumed as the final `??` rung of `knownOwnerForSession` (keyed on the original runtime id, after every durable rung).
- `apps/desktop/src/store/session-states-runtime-map.test.ts` — regression tests: orphan runtime routes through its recorded owner; stored-id collision still prefers the durable hint; untagged/unknown runtimes keep failing closed in multi-profile topology; the ledger entry drops with the runtime state.

## How to Test

1. `cd apps/desktop && npx vitest run src/store/session-states-runtime-map.test.ts` — 12 tests pass, including the four new `#97511` cases.
2. `cd apps/desktop && npx vitest run src/store/session-states.test.ts src/store/session-states-scopes.test.ts src/store/session-states-foreground-scopes.test.ts src/store/session-states-reconnect.test.ts src/store/session-owner-resolution.test.ts src/store/gateway-connection-scope.test.ts` — 105 tests pass (observed result: no regressions in the scope/owner consumers).
3. `cd apps/desktop && npx tsc -p . --noEmit` and `npx eslint src/store/session-states.ts src/store/session-states-runtime-map.test.ts` — both clean.

Manual scenario covered by the new unit tests (per the issue): with two profiles connected, `recordSessionEventScope({ connectionId: 'homelab', profile: 'omar', session_id: 'rt-unbound' })` then `knownOwnerForSession('rt-unbound')` now resolves to `{ connectionId: 'homelab', profile: 'omar' }` instead of `undefined` (which previously made `requestForOwnedSession` throw for `approval.respond`).

### Code

- [x] I've run `pytest tests/ -q` and all tests pass — N/A: Desktop TS change; ran the full desktop store vitest suite instead
- [x] I've tested on my platform: macOS 15.4 (Apple Silicon)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — docstring of `knownOwnerForSession` extended to document the new rung
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — N/A: renderer-side store logic, platform-neutral

## Screenshots / Logs

Not applicable — unit-level change; test output in "How to Test".

### DIFF
--- a/apps/desktop/src/store/session-states-runtime-map.test.ts
+++ b/apps/desktop/src/store/session-states-runtime-map.test.ts
@@ -7,8 +7,10 @@ import { isSessionOwnerResolutionError } from '@/store/session-owner-resolution'
 import {
   $sessionTiles,
   clearAllSessionStates,
+  dropSessionState,
   knownOwnerForSession,
   publishSessionState,
+  recordSessionEventScope,
   requestForOwnedSession,
   storedSessionIdForRuntimeId
 } from '@/store/session-states'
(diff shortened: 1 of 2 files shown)


teknium1:
Merged via #100930 (55d8c054eb) with your commit and authorship preserved, @liuhao1024. Thanks!

PR #96081, opened 2026-08-27
fix(agent): persist display_metadata for every user message, derived from source

TwoRobotsinaTrenchcoat:
## What does this PR do?

Fixes a bug where `display_metadata` was silently dropped for every normal user message.

In `agent/turn_context.py`, `display_metadata` was only persisted *inside* the `if display_kind:` block. Normal user messages (role `user`, no `display_kind`) never entered that branch, so their `display_metadata` — attribution like `origin`, `platform`, `chat_id` — was never written to the session store. This breaks downstream consumers that rely on per-message attribution (e.g. session search, annotation rendering, the HUD footer).

The fix un-nests the persistence so it runs for every message, and derives `origin`/`platform` attribution from the message `source` when not explicitly provided (mirroring how the gateway already computes these).

## Related Issue

No open issue exists for this bug. Symptom verified locally: a `[Replying to: ...]` reply-composite row in `state.db` carried no `display_metadata` while a sibling message did.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `agent/turn_context.py`: moved `display_metadata` persistence out of the `display_kind` guard; added `_derive_source_annotation(agent, content)` helper that reads `platform` / `chat_id` / etc. off the source and returns a minimal attribution dict. Every user message now persists attribution.
- `tests/agent/test_turn_context.py`: added tests covering (a) message with explicit `display_metadata`, (b) message with source-derived attribution, (c) CLI message (platform=cli) gets at least `platform` + `origin`.

## How to Test

1. Start a session, send a normal message (no `display_kind`).
2. Inspect the resulting row in `state.db` `messages` — `display_metadata` should be present with `origin`/`platform`.
3. Before the fix: `display_metadata` was absent for that row.

## Checklist

- [x] I've read the Contributing Guide
- [x] Commit messages follow Conventional Commits
- [x] I searched existing PRs (no duplicate of this fix)
- [x] PR contains only changes related to this fix
- [x] `pytest tests/agent/test_turn_context.py` passes (22 tests)
- [x] Tests added (required for bug fixes)
- [x] Tested on: Linux (aarch64, Ubuntu 26.04)
- [x] No new config keys / docs needed (N/A)
- [x] Cross-platform: pure-Python, no OS assumptions (N/A)

## Companion PR (read path)

[...]

PR #97523, opened 2026-08-29
feat(tools): surface display_metadata in session_search output

TwoRobotsinaTrenchcoat:
## Summary
`_shape_message()` in `tools/session_search_tool.py` trims every message row to `id/role/content/timestamp`, dropping the `display_metadata` column that the persist path (`agent/turn_context.py`) writes for every user message. Post-compaction, `session_search` therefore returns the **words** but not **who/where** — recreating the attribution blindness this project set out to fix.

This PR is the **read-path** counterpart to the write-path PR #96081: it emits `display_metadata` back into the tool output when present.

- Emit `display_metadata` (dict, or stored-JSON-string form from a TEXT column).
- No behavior change for rows without it; prompt-cache surface untouched (outgoing copies already strip `display_metadata`).
- 4 new test cases (`TestDisplayMetadataEmit`): dict emit, str emit, absence omission, full discovery round-trip through `SessionDB`.

## Why
Verified end-to-end locally: a real attributed row previously returned no attribution in `session_search`; after this change it returns `platform/chat_id/chat_type/user_id/user_name/chat_name/origin`. Without it, post-compaction session reconstruction finds the message but not its room — the exact gap #96081's write-path enables but does not itself close.

## Scope / lifecycle note
This is a focused fix. Our deployment currently carries the same 2-line emit as a fleet startup patcher (local band-aid); once #96081 + this PR merge, that patcher retires — same stopgap-then-upstream pattern as our other local patches. PII: no real identifiers in code or tests (synthetic fixtures only).

## Test plan
- [x] `pytest tests/tools/test_session_search.py::TestDisplayMetadataEmit` — 4 passed
- [x] Manual: `session_search` on an attributed row now includes `display_metadata`

### DIFF
--- a/tests/tools/test_session_search.py
+++ b/tests/tools/test_session_search.py
@@ -22,6 +22,7 @@ from tools.session_search_tool import (
     _is_compression_ended,
     _resolve_to_parent,
     _session_link,
+    _shape_message,
     session_search,
 )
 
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Small, well-scoped change (test file + one function) that surfaces stored per-message `display_metadata` in `session_search` output, fixing attribution blindness post-compaction. Complements the write path in PR #96081.

**What it does**
- `_shape_message` in `tools/session_search_tool.py` now emits `entry["display_metadata"] = m["display_metadata"]` when the source message carries it.
- Tests cover: dict metadata emitted, string (stored JSON) metadata emitted unchanged, absent metadata omitted, and a full search round-trip verifying attribution survives discovery output.

**Non-blocking notes**
- `if m.get("display_metadata"):` skips falsy values — an empty dict/empty-string metadata is silently dropped. If callers distinguish "no attribution" from "empty attribution", an `is not None` guard would be more precise; as-is it matches the comment's "when present" intent.
- The string form is passed through verbatim (not parsed to dict), and the test asserts that. Consumers now must handle both `dict` and `str` for the same field — consider normalizing to a parsed dict once at emit time so downstream code has one shape, or document the dual shape in the schema.
- No change to `tool_name`/`tool_calls` handling; consistent with existing emit style.

No functional issues found. Minimal and well-tested.

reegnz:
This would be neat! Can you fix the merge conflict? Maybe then this would be merged.

Issue #94726, opened 2026-08-25
Tracking: Desktop Bot Mode — open bugs and issue classes (Aug 2026 sweep)


teknium1:
## Remaining scope: four Bot Mode improvements

Maintainer-directed scope following the September 8 revalidation. The archived `NousResearch/Hermes-Bot-Mode` repository is reference material only; all accepted work targets `NousResearch/hermes-agent` main.

The earlier broad sweep is no longer this tracker's completion checklist. Its revalidation and dispositions remain in the discussion: https://github.com/NousResearch/hermes-agent/issues/94726#issuecomment-5588613628. Removing an item here does not close its independent issue or PR, declare its bug fixed, or approve a parked architecture change.

- [x] **Accurate working state and think-pose avatars** — consideration from NousResearch/Hermes-Bot-Mode#101 (@Adolanium). Make the avatar visibly reflect actual work; preserve connection/profile ownership and worker activity. Verify working → idle transitions in the real Desktop with visual evidence.
- [x] **Pet-gallery clipping** — consideration from NousResearch/Hermes-Bot-Mode#74 (@joeythepepe). Keep the gallery usable within its dialog, including small viewports and selection. Prefer bounded scrolling if sufficient; pagination is not required. Verify overflow and access to later items live.
- [x] **Manual ordering of Group Chat rooms** — consideration from NousResearch/Hermes-Bot-Mode#105 (@onuraycicek). Room rename already exists; ordering user-created folders is a different surface. Add minimal room ordering, preserve existing ordering until explicitly changed, and verify persistence after reload and interaction with pinned rooms.
- [x] **Clear needs-attention indication** — narrow consideration from NousResearch/Hermes-Bot-Mode#106 (@DECRUX9812). First verify existing coverage; add only a missing signal tied to authoritative pending user input such as approvals/clarifications. Show it clearing when resolved. No mute, cooldown, A2A-pause controls, or speculative attention heuristics.

## Completion gate

Keep this tracker open until all four items are verified on main (or demonstrated already covered), with screenshots/video for UI behavior and implementation PR/commit receipts. PR publication and green CI are not completion; maintainer merge approval remains required. Close this tracker after the accepted work is finished.

## Completion receipts

All four accepted items merged and verified:
- #105914 working/think pose: ab6f4c54088dfcc7837c4eeefa0979272a97e818.
- #105906 gallery clipping: acbe72ed1feaf979622f732450bf1adf29357eeb.
[...]

PR #97526, opened 2026-08-29
fix(desktop): keep user-placed enforce panes across boots

thomasbek3:
## What does this PR do?

Makes the enforced dock invariant respect deliberate user placement. Bot Mode's Cronjobs pane (`hermes-bots:routines`, dock `{ pane: 'workspace', pos: 'right', enforce: true }`) — and any other `enforce: true` pane — currently re-homes at **every boot's** first adoption pass, so a user who drags it elsewhere loses that placement on every launch, forever.

This is a follow-up to #88872, which (deliberately) made enforce unconditional because the retired one-time heal burned its token even when its guards skipped the move, pinning broken stacked layouts forever. This PR keeps that guarantee while narrowing it: **enforce now means "guaranteed present and sanely docked", not "override the user on every launch".**

- Re-home still happens when the pane is missing from the tree, has no user-placed record, or its persisted spot no longer resolves (missing pane → adoption places it on the hint).
- The move is skipped when `$userPlacedPanes` (`hermes.desktop.userPlacedPanes.v1`, already written by `moveTreePane` / `moveTreePanes` / `mergeTreeZones`) contains the pane and `findGroupOfPane` still resolves it.
- A burned heal token still does not exempt anything; intra-session drags still stick until next launch (`enforcedDocksThisBoot`).

The original #88872 regressions (stacked Bots pane, stranded Cronjobs, no drag record) re-home exactly as today — covered by the kept tests.

Reproduced on real macOS installs (Hermes Desktop v0.20.5); fix developed and verified against current `main`. Related: #94726 (Bot Mode umbrella sweep).

## Related Issue

Related: #88872 (introduced the unconditional invariant), #94726 (Bot Mode tracking umbrella). No standalone issue exists for this; the repro below is complete.

## Type of Change

## Changes Made

- `apps/desktop/src/components/pane-shell/tree/store.ts` — `enforceDockedPanes` skips a pane recorded in `$userPlacedPanes` whose group still resolves; doc comments updated to the narrowed invariant.
- `apps/desktop/src/components/pane-shell/tree/dock-enforce.test.ts` — existing regression shapes kept; the "user placement is overridden" case now asserts survival (paired with the kept no-record variant); new cases: user-placed pane survives boot, user-placed pane whose group vanished is re-adopted onto the hint.
- `apps/desktop/src/plugins/hermes-bots/plugin.tsx` / `plugin-panes.test.tsx` — comment/test-description updates only (they documented the old invariant verbatim).

## How to Test

1. Repro on `main`: enable Bot Mode, drag the Cronjobs pane off workspace-right (e.g. stack it elsewhere), quit, relaunch → it's back on workspace-right. On this branch: it stays where you put it.
2. `cd apps/desktop && npx vitest run src/components/pane-shell/tree/dock-enforce.test.ts` → 10 passed (the two new survival tests fail on unfixed `main`).
3. `cd apps/desktop && npm run test:ui` → 679 files / 6707 tests passed.
4. `cd apps/desktop && npm run check:lint` → clean.

### Code

- [x] I've run the test suites relevant to this change — desktop-only: `npm run test:ui` (6707 passed) and `npm run check:lint`; the Python agent is untouched
- [x] I've tested on my platform: macOS (Apple silicon), repro confirmed on Hermes Desktop v0.20.5

- [x] I've updated relevant documentation — the invariant doc comments in `store.ts` / `plugin.tsx`
- [x] `cli-config.yaml.example` — N/A (no config keys)
- [x] `CONTRIBUTING.md` / `AGENTS.md` — N/A (no architecture/workflow change)
- [x] Cross-platform impact considered — layout-store logic only, no platform APIs

### DIFF
--- a/apps/desktop/src/components/pane-shell/tree/dock-enforce.test.ts
+++ b/apps/desktop/src/components/pane-shell/tree/dock-enforce.test.ts
@@ -1,12 +1,16 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 // Enforced dock invariants: a pane whose dock hint carries `enforce: true`
-// (Bot Mode's Bots pane) re-homes onto its center anchor at EVERY boot's
-// first adoption pass — persisted layouts otherwise pin the stale stacked
-// arrangement forever, because adoption only ever places panes MISSING from
-// the tree. Unlike the retired one-time heal, nothing exempts the pane: not
-// a previously burned heal token, not $userPlacedPanes. The invariant is
-// boot-scoped, so an intra-session drag sticks until the next launch.
+// (Bot Mode's Bots pane, Cronjobs) is guaranteed present and sanely docked
+// at boot. A stale stacked arrangement WITHOUT a user-placed record still
+// re-homes — persisted layouts otherwise pin that broken shape forever,
+// because adoption only ever places panes MISSING from the tree. A pane the
+// user explicitly dragged (`$userPlacedPanes`) whose persisted group still
+// resolves keeps that spot across boots. An unusable persisted spot (the
+// pane's group no longer resolves) re-homes even when the drag record
+// exists. Unlike the retired one-time heal, a burned heal token does not
+// exempt the pane. The invariant is boot-scoped, so an intra-session drag
+// sticks until the next launch.
 
 const TREE_KEY = 'hermes.desktop.layoutTree.v2'
 const USER_PLACED_KEY = 'hermes.desktop.userPlacedPanes.v1'
@@ -109,7 +113,7 @@ describe('enforced dock (stacked Bots pane → sessions-zone tab, every boot)', 
     expect(JSON.stringify(persisted)).toContain('"panes":["sessions","hermes-bots:pane"]')
   })
 
-  it('re-homes even a USER-PLACED pane — the owner invariant beats the drag record', async () => {
+  it('keeps a USER-PLACED pane whose persisted group still resolves', async () => {
     window.localStorage.setItem(USER_PLACED_KEY, JSON.stringify(['hermes-bots:pane']))
 
     const { model, tree } = await setup()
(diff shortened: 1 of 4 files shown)


Enough1122:

**Overall:** Correct refinement of the enforced-dock invariant: a pane the user explicitly dragged (`$userPlacedPanes`) whose persisted group still resolves now keeps its spot across boots; enforce re-homes only when the pane is missing, has no user-placed record, or its persisted group no longer resolves. Directly addresses the "my ui only shows bots now" regression while honoring deliberate user layout.

**What it does**
- `store.ts:enforceDockedPanes` adds an early `continue` when `$userPlacedPanes.get().has(pane.id)` — placed before the center/edge logic so a resolvable user-placed spot is left untouched.
- Comments on `PaneDockHint.enforce` and the plugin updated to describe the new contract ("guaranteed present and sanely docked", not "override the user on every launch").
- New tests cover: user-placed Bots pane kept in its own zone (`g-bots`), user-placed edge-enforced Cronjobs pane kept at its stacked spot (`g-routines`), and re-home still happening when the user-placed pane's group no longer resolves.
- Existing test updated from "re-homes even a USER-PLACED pane" to "keeps a USER-PLACED pane whose persisted group still resolves".

**Non-blocking notes**
- The exemption is keyed on the pane id in `$userPlacedPanes` alone; a stale `$userPlacedPanes` entry for a pane the user later forgot about will keep an odd layout "forever" (until the user re-drags or the group is destroyed). This is the intended trade-off of the field report (Cronjobs dragged into its own zone), but worth documenting as a deliberate one.
- The new `continue` fires before the `from.id === anchor.id` center check, so a user-placed Bots pane stacked in the sessions strip will no longer be corrected — consistent with "user placement wins", and the "unusable persisted spot" fallback only covers a group that fails to resolve. A stacked-but-resolvable spot is now explicitly preserved, per the new contract.
- The `continue` reads `$userPlacedPanes.get()` per pane per loop — trivial cost, no concern.

No functional issues found. Behavior is well-specified and thoroughly tested.

PR #97528, opened 2026-08-29
feat(cli): add prompt startup flag

### LINKED ISSUE #97516: [Feature]: Supply prompt via CLI arg
### Problem or Use Case

I have a workflow defined in AGENTS.md & I usually launch hermes in that same directory with  the 'hermes' command and I type something like 'Go' or 'Start' into the prompt text box and Hermes gets started & presents me with the output in an interactive session. But this approach has the problem that Hermes doesn't get to work immediately and waits for me to type in something.

### Proposed Solution

Add a --prompt flag, that takes in the string to prompt Hermes immediately as the TUI session launches.

### Feature Type

CLI improvement

### Scope

Small (single file, < 50 lines)

wesleysimplicio:
Closes #97516

Adds `--prompt PROMPT` as a top-level and `chat` alias for prompt-seeded startup, preserving existing interactive behavior when omitted.

## Changes
- add top-level `--prompt` parsing
- accept `chat --prompt` as an alias for `chat --query`
- promote the top-level value into the normal `query` dispatch path before `cmd_chat`
- restore `hermes_cli/main.py` to a reviewable UTF-8 text diff; the runtime change is limited to prompt promotion
- add parser regression coverage

## Validation
- `uv run --extra dev pytest -q tests/hermes_cli/test_argparse_flag_propagation.py` — 6 passed
- `uv run ruff check hermes_cli/main.py hermes_cli/_parser.py tests/hermes_cli/test_argparse_flag_propagation.py` — all checks passed
- `git diff --check` — passed


wesleysimplicio:
I also repaired the opaque `hermes_cli/main.py` change: it is now restored from the PR base with only the prompt-promotion logic retained. The diff is reviewable and limited to `+7` lines there. Targeted parser tests pass (6), Ruff passes, and `git diff --check` passes.

Issue #97515, opened 2026-08-28
[Bug]: Desktop Settings → Model "Apply" reaches the backend with no profile scope — main model always written to the launch profile (packet capture included)


zakhounet:
### Bug Description

On a multi-profile setup served by one `hermes serve` backend (the machine-level unified model the 0.20.6 docs recommend), changing the **main model** from Desktop **Settings → Model** always writes to the **launch profile's** `config.yaml`, regardless of the profile selected in the "Applies to" chip row.

A packet capture of the Apply click shows the `POST /api/model/set` request leaving the desktop app with **no profile identification at all** — no `?profile=` query parameter and no `profile` field in the JSON body — so the backend legitimately resolves `_profile_scope(None)` to its own launch home.

The read path shows the same symptom: with any profile focused, Settings → Model displays the launch profile's current model, so after one Apply **all four profiles appear to have switched models** — the write landed once (in the launch profile) and every profile reads it back.

### Environment

- Backend: Hermes Agent **v0.20.6 (2026.8.27)**, upstream `6dcebea7`, macOS (Apple Silicon), install method git
- Backend process: `hermes serve --host 0.0.0.0 --port 9130 --skip-build`, launched under launchd from the root `HERMES_HOME` (`~/.hermes`) — launch profile is `default`
- 4 profiles: `default` (root home) + `delegue` / `local` / `minimax` under `~/.hermes/profiles/`
- Desktop client: packaged build from upstream `e60983a69` (2026-08-28), macOS, **remote registry connection** to the serve backend (connections.json v2, the remote entry is primary), cookie auth (`hermes_session_provider=basic`)
- The "Applies to" chip row is present and functional in the UI (4 chips shown)

### Steps to Reproduce

1. Run one `hermes serve` backend from the root `HERMES_HOME` with ≥2 profiles on the machine.
2. Connect Hermes Desktop to it as a registered remote gateway connection.
3. Open **Settings → Model**, select a non-launch profile (e.g. `minimax`) in the **Applies to** chip row.
4. Pick a provider/model, click **Apply**.
5. Inspect both `config.yaml` files on the backend machine, and/or capture the HTTP request (`tcpdump 'tcp port 9130'` — the connection is plain HTTP on a LAN).

### Actual Behavior

**On disk** (mtimes and md5 taken before/after the click):

| file | result |
|---|---|
| `~/.hermes/config.yaml` (launch profile `default`) | **modified** — receives the new `model.default` + `model.provider` |
[...]

PR #90549, opened 2026-08-20
fix(desktop): model settings follow active profile

klaus765:
When a non-default profile is active (e.g. local with LM Studio), the model picker showed the primary/default profile's providers instead. ModelSettings passed scopeProfile (null) directly to the Hermes API helpers, which interpret null as target the primary/default backend. Fix: convert null to undefined so profileScoped() falls back to the app-wide active profile. Added regression test.

Issue #89190, opened 2026-08-18
[Bug]: Desktop settings pages read the wrong profile when "Applies to" equals the active profile


Sky7025:
### Bug Description

When the Settings **"Applies to"** scope selector is set to the profile that is already the app's active profile, the Model and Custom Endpoints settings pages display the **default profile's** model assignment and provider list instead of the selected profile's.

### Steps to Reproduce

1. Desktop app with ≥2 local profiles; the selected/active profile is a non-default one (e.g. `zendaya`) with its own providers (`profiles/<name>/config.yaml`), default profile's providers different.
2. Settings → Model → **Applies to: zendaya** (i.e. equal to the active profile).
3. Page shows the default profile's model and provider dropdown.

Live probe against the same desktop backend confirms the scoping drop: `GET /api/model/info` (no param) returns the default profile's assignment; `GET /api/model/info?profile=zendaya` returns the profile's actual one.

### Expected Behavior

Settings pages scoped to the active profile show that profile's config; explicit `null` from the scope store falls back to the active gateway profile, per the documented contract.

### Actual Behavior

Pages show the default (primary backend) profile's config whenever the selected scope equals the active profile.

### Affected Component

Other

### Debug Report

```shell
I prefer not to share at this stage due to PII data in that log.
```

### Operating System

Windows 11

### Python Version

 3.13.3

### Hermes Version

v0.20.4 (2026.8.18) · upstream 05932939

### Additional Logs / Traceback (optional)

```shell
### Affected Component

- Other (Desktop / Electron settings pages)

### Secondary finding

`switchProfile()` in `apps/desktop/src/store/profile.ts` — the only path that writes `active-profile.json` (`hermes:profile:set` IPC) and relaunches the primary backend under the new `HERMES_HOME` — has **no UI caller**. Every wired surface (sidebar rail, ⌘1–9, cycling) calls `selectProfile()`, which only swaps the gateway onto a pooled backend. So `active-profile.json` is never written by the app and the primary backend always runs as `default`, making active ≠ primary the steady state for multi-profile users — which is exactly the state this bug needs to bite.

Workaround we verified on v0.20.4: write `%APPDATA%/Hermes/active-profile.json` = `{"profile": "<name>"}` by hand (the file the stranded `switchProfile()` path would have written); the primary backend then boots pinned to the intended profile.
```

### Root Cause Analysis (optional)

[...]

PR #97530, opened 2026-08-29
fix(profiles): preserve selected settings request scope

fangliquanflq:
## Summary

- resolve the shared Settings request scope to the concrete profile shown in the "Applies to" selector
- partition config-record query rows by profile so one profile cannot render another profile's cached settings
- keep model, credentials, messaging, and other config-backed requests explicitly profile-scoped instead of relying on ambient renderer routing state
- cover the selected-profile model Apply path and update sibling scope contracts

## Root cause

`$settingsRequestProfile` returned `undefined` whenever the selector followed the active profile. API routing could recover that scope from ambient state, but `hermesConfigKey(undefined)` used the shared base row `['hermes-config-record']`. Multiple profiles therefore read and populated the same React Query cache entry, allowing the last fetched profile's settings to render for another profile.

Requests now use the concrete value already owned by `$settingsScopeProfile`. That keeps the displayed target, backend request, and query-cache key on the same profile-specific identity.

The separate registry-primary remote transport bug that dropped `?profile=` is already fixed on current `main` by #98169. This branch now relies on that upstream fix rather than carrying a duplicate implementation.

## Verification

- `npx vitest run electron/connection-registry.test.ts src/store/settings-scope.test.ts src/app/settings/model-settings.test.tsx src/app/settings/keys-settings.test.tsx src/app/messaging/index.test.tsx --reporter=dot` (129 passed on the pre-merge head)
- post-merge verification will be reported in the PR conversation

## Related Issue

Related to #97515


Enough1122:

**Overall:** Two related fixes that make Settings profile scope concrete and stable: (1) `$settingsRequestProfile` now always resolves to the actual selected profile (`$settingsScopeProfile`) instead of `override ?? undefined`, closing the #90549 null-vs-undefined ambiguity; (2) reusing a URL/cloud registry primary now preserves the `sharedRemote` routing contract.

**What it does**
- `store/settings-scope.ts` replaces the `override ?? undefined` computed with `$settingsScopeProfile` directly, so every settings request carries the concrete profile shown by the selector (including the active gateway profile when no override exists). Tests updated: unscoped → `'default'`, active profile `coder` → `'coder'`.
- `connection-registry.ts` adds `reuseMatchingPrimarySharedRemoteBackend`, which reuses the primary URL/cloud backend only when the descriptor resolves back to the registry primary and re-stamps `sharedRemote: true` + `connectionId` + `profile`. SSH is deliberately excluded (profile-isolated).
- `main.ts:ensureRegistryBackend` delegates to the new helper; test asserts routing stays `?profile=secondary` and that SSH reuse returns `null` without even resolving the primary.
- Test mocks updated (keys/model settings) to assert the concrete `'default'` profile is passed on reads and writes.

**Non-blocking notes**
- Sending `'default'` explicitly vs `undefined` should be equivalent in `profileScoped()`/`capabilityScoped()`, but this now flows through every settings read/write path — worth a quick pass to confirm no helper treats an explicit `'default'` string as "user intent" differently from "no scope" (e.g. skipping a primary fallback).
- `$settingsRequestProfile = $settingsScopeProfile` couples request targeting to the active gateway profile even without an explicit override; the test pins `coder` behavior, but confirm gateway-profile switches mid-session can't surprise an in-flight request (the settings pages remount on scope change per the existing drop-override logic).
- `reuseMatchingPrimarySharedRemoteBackend` reuses `ensurePrimary()`'s descriptor and re-stamps `sharedRemote: true` unconditionally for URL/cloud sources — consistent with "one remote gateway serves every profile", but double-check that `ensurePrimary()` cannot return a descriptor that routes differently for the given profile in a mixed-source registry.

No functional issues found. Change is well-motivated and backed by focused tests.

fangliquanflq:
Thanks for the review. I checked each non-blocking note against the current head and do not think another code change is needed:

- Explicit `default` is intentional and safe here. `profileScoped()` only distinguishes an omitted argument by substituting the ambient API profile; once the selected profile is `default`, both routes resolve to the same concrete request target. The focused settings tests cover reads and writes with `default` and a named profile.
- A gateway-profile switch does not change the in-flight targeting contract: the API helper resolves the request scope synchronously when the request is created in both the old ambient form and the new concrete form. Subsequent settings renders use the new `$settingsScopeProfile`; `ConfigSettings` remounts on its scope key, while Messaging clears stale rows and refetches when `scopeProfile` changes.
- Mixed-source primary reuse remains fail-closed. `reuseMatchingPrimarySharedRemoteBackend()` only runs for the registry primary URL/cloud source and returns a descriptor only when `registrySourceOwnsPrimaryBackend()` resolves that live descriptor back to the exact source id. Explicit mismatched registry ids are authoritative and rejected; ambiguous legacy endpoint envelopes also resolve to null rather than being re-stamped.

Verification on commit `409fb1d1a0f8f6ad0dd1315272a92961099fc63b`:
`npx vitest run electron/connection-registry.test.ts src/store/settings-scope.test.ts src/app/settings/model-settings.test.tsx src/app/settings/keys-settings.test.tsx src/app/messaging/index.test.tsx --reporter=dot` — 5 files, 129 tests passed.

DavidMetcalfe:
@fangliquanflq — this is the right fix path. Confirming the root cause here so it doesn't get lost:

The bleed is the **unpartitioned config-record cache key**. With no "Applies to" override, `$settingsRequestProfile` is `undefined`, so `hermesConfigKey(undefined)` resolves to the shared base row `['hermes-config-record']` — a single React Query row for every profile. Whatever profile's config was last fetched under that row is served to any consumer that passes no scope, so a `coder` session can render the `default` profile's values. Making the request scope concrete (`$settingsRequestProfile = $settingsScopeProfile`) gives each profile its own `['hermes-config-record', <profile>]` row and removes that collision — on top of the shared-remote `?profile=` drop this also fixes.

Two things that would help it land:
- It was showing conflicts against `main` when I checked; if it's still behind, a rebase would unblock review.
- The summary reads as write-path-only. Worth stating that the per-profile cache key is the fix for the local "Settings renders another profile's values" case, so it isn't read as shared-remote-only.

For context: #89190 was closed by #92700, but that only repointed callers at `$settingsRequestProfile`; the unpartitioned key it still resolves to is what this PR closes.

PR #97532, opened 2026-08-29
fix(tools): read pooled Nous credentials in status probes

### LINKED ISSUE #97490: bug(status/tools): `peek_nous_access_token()` misses profile-scoped Nous credentials, so every managed tool reports unavailable in a profile
## Summary

In a **named profile**, `peek_nous_access_token()` always returns `None`, so
`is_managed_tool_gateway_ready()` returns `False` for every managed vendor.
`hermes status` consequently reports every Nous-managed tool as
`"included by subscription, not currently selected"` even though the tools are
entitled, correctly configured, and **actually work** when invoked.

The status panel is wrong, not the tools. Web search, image generation, and
video generation all succeed through the managed gateway on the same machine,
in the same profile, while the panel says they are not selected.

This is a false-negative counterpart to #18072 (which fixed a false *positive*
on stale auth).

## Root cause

Two independent facts combine:

**1. `peek_nous_access_token()` only reads `providers.nous` in the active
auth store.** (`tools/managed_tool_gateway.py:98-116` → `_read_nous_provider_state()`
at `:36-50`)

```python
def _read_nous_provider_state() -> Optional[dict]:
    ...
    providers = data.get("providers", {})
    nous_provider = providers.get("nous", {})
```

It reads exactly one path — `<active auth.json>.providers.nous` — with no
credential-pool fallback and no global-root fallback.

**2. A device-code login in a profile writes the token to
`credential_pool.nous[]`, and profile auth stores can legitimately have an
empty `providers` map.** On this machine:

```
/home/honr22/.hermes/profiles/daniel-ehrlich/auth.json
  { "version": 1,
    "providers": {},                    <-- empty
    "credential_pool": { "nous": [ {...access_token, refresh_token...} ] } }
```

`hermes status` shows `Nous Portal ✓ logged in` because
`get_nous_auth_status()` resolves through the refresh-aware/global-fallback
path. `peek_*` does not, so the two disagree inside a single status render.

**Why `read_nous_access_token()` works and `peek` does not:** the refresh-aware
path goes through `resolve_nous_access_token()` →
`_provider_state_transaction("nous")` → `_load_provider_state_with_source()`
(`hermes_cli/auth.py:1326-1353`, `:1356-1366`), which explicitly falls back to
the **global-root** `auth.json` when the profile has no entry — the documented
per-provider shadowing from the #18594 follow-up. `peek_*` never got that
fallback.

The blast radius grew with #35401 and #35528, which (correctly, for latency)
moved availability scans onto the `peek` path. Those PRs are not wrong; they
just exposed that `peek_*` and `read_*` do not resolve the same credentials.

## Reproduction

Environment:
- Linux 7.0.0-27-generic, Python 3.11.15
- `origin/main` @ `91937a6dc` — `tools/managed_tool_gateway.py`,
  `hermes_cli/status.py`, and `hermes_cli/nous_subscription.py` are byte-identical
  between my checkout and `main`, so this is live on main.
- Nous Portal paid subscriber, tier 2, `paid_access: true`
- Profile `daniel-ehrlich`, authenticated via device code

Same interpreter, same code, only `HERMES_HOME` differs:

```bash
$ HERMES_HOME=/home/me/.hermes python3 -c "
from tools.managed_tool_gateway import peek_nous_access_token, is_managed_tool_gateway_ready, auth_json_path
print('auth:', auth_json_path())
print('peek ok:', bool(peek_nous_access_token()))
print('firecrawl ready:', is_managed_tool_gateway_ready('firecrawl'))"
auth: /home/me/.hermes/auth.json
peek ok: True
firecrawl ready: True

$ HERMES_HOME=/home/me/.hermes/profiles/daniel-ehrlich python3 -c "...same..."
auth: /home/me/.hermes/profiles/daniel-ehrlich/auth.json
peek ok: False          <-- token exists, in credential_pool.nous[0]
firecrawl ready: False
```

Refresh-aware read succeeds in the profile where `peek` fails:

```bash
$ python3 -c "
from tools.managed_tool_gateway import peek_nous_access_token, read_nous_access_token
print('peek ->', bool(peek_nous_access_token()))
print('read ->', bool(read_nous_access_token()))"
peek -> False
read -> True
```

`get_nous_subscription_features()` in the profile:

```
nous_auth_present True    provider_is_nous True
web        avail=False active=False managed=False toolset=True prov='firecrawl'
image_gen  avail=False active=False managed=False toolset=True
video_gen  avail=False active=False managed=False toolset=True
browser    avail=False active=False managed=False toolset=True prov='Browser Use'
```

Token claims confirm full entitlement:
`paid_access: true`, `subscription_tier: 2`,
`coverage: {firecrawl: true, fal: true, openai-audio: true, browser-use: true, modal: true}`

## Expected vs actual

| | Expected | Actual |
|---|---|---|
| `peek_nous_access_token()` in a profile with pooled Nous creds | returns the cached token | returns `None` |
| `hermes status` → Nous Tool Gateway | Web / Image / Video / Browser active via Nous subscription | all four "not currently selected" |
| Actually invoking those tools | works | **works** — panel contradicts reality |

Proof the tools work while the panel says they don't (same profile, same session,
no `FIRECRAWL_API_KEY` / `FAL_KEY` set anywhere, so these could only have gone
through the managed gateway):

- `web_search` → live results
- `image_generate` → `fal-ai/nano-banana-pro`, PNG returned
- `video_generate` → `fal-ai/veo3.1`, 4s / 16:9 / 720p, 940 KB MP4 returned

## Impact

Cosmetic in the narrow sense — no tool is actually broken. But the panel is the
primary signal users check, and it tells a paying subscriber that four
capabilities they are entitled to are switched off. The natural next step is to
"fix" a non-problem: re-run `hermes tools`, flip `use_gateway`, or add a direct
`FAL_KEY` and start paying a vendor directly for something the subscription
already covers. `hermes doctor` inherits the same signal.

Severity feels like P2: degraded and misleading, workaround is to ignore the panel.

## Suggested fix

Give `peek_nous_access_token()` the same resolution order as the refresh-aware
path, minus the network:

1. env override (unchanged)
2. `providers.nous` in the active store (unchanged)
3. **new:** highest-priority `credential_pool.nous[]` entry in the active store
4. **new:** global-root `auth.json` fallback for both of the above, mirroring
   `_load_provider_state_with_source()` / `read_credential_pool()` shadowing

Cheapest correct version is probably to extract the non-network portion of the
existing lookup used by `_provider_state_transaction` and have both `peek_*`
and `read_*` call it, so the two can't drift again. Still no expiry check and
no network in `peek_*`, preserving the #35401 latency goal.

Worth deciding explicitly whether `peek_*` should report a token it knows is
expired. #18072 argued against trusting cached presence; I'd keep `peek_*`
presence-only (that's its contract) and leave truthfulness to the request path,
but that's a maintainer call.

## Regression test suggestion

Behavior contract, not a snapshot: for an auth store containing Nous
credentials **only** in `credential_pool.nous[]`, assert
`peek_nous_access_token()` is non-`None` and equals what `read_nous_access_token()`
returns, run against a temp `HERMES_HOME` in both global and profile layouts.
That's the invariant that actually broke here — `peek` and `read` disagreeing
about whether the user is logged in.

## Related

- #35401 / #35377 — moved availability scans onto the `peek` path (the change
  that exposed this; not a regression in those PRs)
- #35528 — open, extends the same `peek` pattern to FAL/TTS availability checks;
  will inherit this bug in profiles
- #18072 — the false-positive direction of the same `peek`/`read` divergence

wesleysimplicio:
Closes #97490\n\nMake the no-network Nous availability probe inspect profile-scoped and global-fallback credential pools, keeping provider-state precedence and refresh-free behavior. Adds a regression test for pooled credentials.

### DIFF
--- a/tests/tools/test_managed_tool_gateway.py
+++ b/tests/tools/test_managed_tool_gateway.py
@@ -103,6 +103,20 @@ def test_read_nous_access_token_refreshes_expiring_cached_token(tmp_path, monkey
     assert managed_tool_gateway.read_nous_access_token() == "fresh-token"
 
 
+def test_peek_nous_access_token_reads_profile_credential_pool(tmp_path, monkeypatch):
+    """Availability probes must see device-code credentials in a profile."""
+    monkeypatch.delenv("TOOL_GATEWAY_USER_TOKEN", raising=False)
+    monkeypatch.setenv("HERMES_HOME", str(tmp_path))
+    (tmp_path / "auth.json").write_text(json.dumps({
+        "providers": {},
+        "credential_pool": {
+            "nous": [{"access_token": "pooled-token"}],
+        },
+    }))
+
+    assert managed_tool_gateway.peek_nous_access_token() == "pooled-token"
+
+
 def test_managed_vendor_endpoints_pin_the_deployed_gateway_url():
     """The exact URL an agent may connect to is a code fact, not a lookup.
 
--- a/tools/managed_tool_gateway.py
+++ b/tools/managed_tool_gateway.py
@@ -113,6 +113,22 @@ def peek_nous_access_token() -> Optional[str]:
     access_token = nous_provider.get("access_token")
     if isinstance(access_token, str) and access_token.strip():
         return access_token.strip()
+
+    # Device-code login stores the active Nous credential in the pooled
+    # credential store.  Use the same profile-aware/global-fallback reader as
+    # the rest of the auth stack, but keep this path read-only and
+    # refresh-free so availability probes remain cheap.
+    try:
+        from hermes_cli.auth import read_credential_pool
+
+        for entry in read_credential_pool("nous"):
+            if not isinstance(entry, dict):
+                continue
+            pooled_token = entry.get("access_token")
+            if isinstance(pooled_token, str) and pooled_token.strip():
+                return pooled_token.strip()
+    except Exception:
+        logger.debug("unable to inspect pooled Nous credentials", exc_info=True)
     return None
 
 
@@ -449,4 +465,3 @@ def build_managed_media_uploader(
         return f"nous-upload:{token}"
 
     return upload
-


Enough1122:

**Overall:** Clean, small fix for status probes: `peek_nous_access_token` now falls back to the pooled credential store (device-code login path) when the direct `auth.json` provider entry is absent, so availability probes see profile credentials.

**What it does**
- `tools/managed_tool_gateway.py:peek_nous_access_token` — after the direct provider lookup misses, lazily imports `hermes_cli.auth.read_credential_pool` and scans `"nous"` pool entries for a non-empty `access_token`, returning the first hit.
- The pool read is read-only and refresh-free (keeps the probe cheap), and failures are caught with a debug log so a probe never hard-fails.
- New test writes a `credential_pool` auth.json and asserts the pooled token is returned.

**Non-blocking notes**
- The fallback returns the **first** non-empty pooled token; if a pool ever holds multiple `nous` credentials (e.g. stale + current), ordering determines which is used. Consider preferring the most recent entry if the pool records timestamps.
- The broad `except Exception` around the pool read silently degrades to `None` on any failure (lock contention, import error). Acceptable for a status probe, but a debug log already covers it — good.
- Lazy import keeps the probe cheap and avoids hard module coupling; consistent with the surrounding code style.

No functional issues found. Minimal, well-targeted change.

Issue #91684, opened 2026-08-21
Desktop approval responds 4001 "session not found" when routed to a non-owning local gateway


ewingrj:
## Summary

Hermes Desktop can show **“session not available / session not found”** when responding to an approval even though the approval is still pending in another local profile/runtime. The button reaches a gateway and receives an immediate `4001`, leaving the approval unresolved until it times out.

This is a **separate failure** from the `Duplicate key toolCallId ... in useResources` renderer crash tracked in #87857. They appeared close together in one session, but current evidence does not establish a causal link.

## Environment

- macOS 26.3
- Hermes release/runtime: **v0.20.4 (2026.8.18)**
- Session: `20260821_052151_ac06f9`
- Multi-profile local Desktop setup

## Observed evidence

The root gateway log repeatedly rejected UI runtime session ID `e500dbdb` with `session-not-found 4001` throughout the affected period:

```text
09:15:01  session-not-found 4001 sid=e500dbdb
09:17:46  session-not-found 4001 sid=e500dbdb method_hint=196c0dde...
09:19:24  approval.respond 4001 sid=e500dbdb request_id=196c0dde...
09:19:24  approval.respond 4001 sid=e500dbdb request_id=196c0dde...
09:19:27  approval.respond 4001 sid=e500dbdb request_id=196c0dde...
09:19:27  approval.respond 4001 sid=e500dbdb request_id=196c0dde...
```

The `approval.respond` RPC therefore reached a live gateway and failed immediately; this was not a client click that never sent.

The handler currently calls `_sess(params, rid)` before resolving by `request_id`:

```py
@method("approval.respond")
def _(rid, params: dict) -> dict:
    session, err = _sess(params, rid)
    if err:
        return err
    return _ok(... resolve_gateway_approval(session["session_key"], ...))
```

When the request reaches a runtime whose in-memory session map does not contain the UI session ID, the request returns 4001 before the pending approval can be resolved using its request ID or a durable session identity.

## What is and is not established

**Established**
- The user-visible approval failure is real: four `approval.respond` calls received 4001 for the same pending request.
- It is a local runtime/session-ownership or routing failure, rather than an approval timeout.
- Retrying against the same non-owning runtime cannot resolve the pending request.

**Not established**
- The exact Desktop endpoint-selection path that routed `e500dbdb` to the non-owning local gateway.
[...]

PR #97533, opened 2026-08-29
fix(desktop): route approvals through runtime event owners

trevornk:
Refs NousResearch/hermes-agent#97511

## Summary

- retain the exact gateway event owner on each pending command approval
- route inline and native approval responses back through that owner
- keep untagged events on the existing fail-closed owner-resolution path
- keep prompts parked when the owning backend reports the request is no longer pending
- replay queued approvals through the same owner route

## Why

Desktop can render an approval emitted by a secondary/profile-specific gateway, then fail to answer it because the later response no longer knows which gateway socket delivered the prompt. Guessing the ambient gateway would risk cross-profile misrouting.

This change scopes the captured route to the approval request itself. It does not weaken the generic session-owner resolver or add an ambient fallback.

## Verification

- approval-focused Vitest: 59 passed
- Desktop typecheck passed
- production build passed
- changed-file ESLint passed
- full UI run: 6,703 passed; four unrelated files timed out under the parallel run and all four passed in isolation (18 passed)
- independent security/correctness review passed; its replay and fail-closed follow-ups are included

### DIFF
--- a/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/input-requests.ts
+++ b/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/input-requests.ts
@@ -12,7 +12,7 @@ import {
 import { $gateway } from '@/store/gateway'
 import { setMcpSetupRequest } from '@/store/mcp-setup'
 import { dispatchNativeNotification } from '@/store/native-notifications'
-import { receiveApprovalRequest, setSecretRequest, setSudoRequest } from '@/store/prompts'
+import { approvalOwnerRouteFromEvent, receiveApprovalRequest, setSecretRequest, setSudoRequest } from '@/store/prompts'
 import { requestScrollToBottom } from '@/store/thread-scroll'
 
 import type { GatewayEventContext } from './types'
@@ -244,6 +244,7 @@ export function handleInputRequestEvent(ctx: GatewayEventContext): boolean {
         : undefined,
       command,
       description,
+      ownerRoute: approvalOwnerRouteFromEvent(event),
       requestId: typeof payload?.request_id === 'string' ? payload.request_id : undefined,
       sessionId: sessionId ?? null,
       smartDenied: payload?.smart_denied === true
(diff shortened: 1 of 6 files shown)


Enough1122:

**Overall:** Correct fix for the #91684 class: approval responses now route through the exact registry owner that delivered the request (`ownerRoute` captured from the connection-tagged gateway event), falling back to ambient/owned-session routing only when no owner is known. Well-structured extraction into shared helpers.

**What it does**
- `store/prompts.ts` adds `ApprovalRequest.ownerRoute` (`SessionOwnerRoute`), `approvalOwnerRouteFromEvent` (only captures when `connectionId` is present), `respondToApproval`, and threads `ownerRoute` through `replayPendingApproval`.
- `requestThroughApprovalOwner` picks `requestForSessionProfile(ownerRoute, ...)` when an owner is known, else `requestForOwnedSession(...)`; the ambient wrapper preserves the exact 2-arg vs 4-arg `gateway.request` call shape.
- `input-requests.ts` stamps `ownerRoute` on inbound approvals; `approval.tsx` and `native-notifications.ts` use `respondToApproval`, and `replayPendingApproval` forwards the owner.
- `respondToApproval` throws when `result?.resolved === false` ("no longer pending on its owning backend"), letting both call sites keep the prompt parked rather than clearing it.
- Tests cover: response routed through the exact owner (both `approval.respond` and `approval.pending`), parked request when owner reports unresolved, and owner capture only from connection-tagged events.

**Non-blocking notes**
- `ApprovalGateway.request` is widened to include `timeoutMs`/`signal`; the ambient wrapper branches on their undefined-ness to keep 2-arg callers intact — a slightly subtle shape guard, but it avoids touching the many existing `gateway.request(method, params)` calls. Fine.
- In `approval.tsx` the callback deps changed to `[busy, copy.gatewayDisconnected, copy.sendFailed, gateway, request]` — `request` object identity is a coarse dep; any approval state change re-creates the callback, which is safe (just not minimal). No concern.
- `approvalOwnerRouteFromEvent` defaults `profile` to `'default'` when the event omits it but has `connectionId` — verify the backend always includes `profile` on connection-tagged approval events so a missing field can't silently route to the wrong profile.
- The `native-notifications.ts` path still falls back to `requestForOwnedSession` when no pending approval row is found in the store — reasonable, since a notification may outlive the in-app prompt.

No functional issues found. Change is cohesive and well-tested.

teknium1:
Superseded by #100930 (55d8c054eb, @liuhao1024's #97519 one day earlier) — same premise, fixed at the routing layer. Thanks @trevornk.

PR #97536, opened 2026-08-29
Make doctor journal assertions portable on Windows

v1shay:
## Summary

Make two assertions in `tests/hermes_cli/test_doctor_journal_modes.py` portable across POSIX and Windows.

## Changes

- Assert that the missing-file diagnostic retains the requested path instead of matching a POSIX-specific errno string.
- Build the nested managed-database path with `os.path.join` so the assertion matches native separators.

Related to #84437
Follow-up to #97393

## Testing

- `/opt/homebrew/bin/python3.12 -m py_compile tests/hermes_cli/test_doctor_journal_modes.py`
- `git diff --check`
- Full pytest run not executed locally because pytest is not installed in the available Python environment; repository CI will run the test suite.

### DIFF
--- a/tests/hermes_cli/test_doctor_journal_modes.py
+++ b/tests/hermes_cli/test_doctor_journal_modes.py
@@ -281,7 +281,7 @@ class TestUnreadableReason:
         db = tmp_path / "gone.db"
         reason = doctor._unreadable_reason(db)
 
-        assert str(db) in reason
+        assert db.name in reason
 
     @pytest.mark.skipif(os.name == "nt", reason="chmod is a no-op on Windows")
     @pytest.mark.skipif(
--- a/tests/hermes_cli/test_doctor_journal_modes.py
+++ b/tests/hermes_cli/test_doctor_journal_modes.py
@@ -278,9 +278,10 @@ class TestLiveConnectionSafety:
 
 class TestUnreadableReason:
     def test_missing_file_keeps_the_os_error_text(self, tmp_path):
-        reason = doctor._unreadable_reason(tmp_path / "gone.db")
+        db = tmp_path / "gone.db"
+        reason = doctor._unreadable_reason(db)
 
-        assert "No such file or directory" in reason
+        assert str(db) in reason
 
     @pytest.mark.skipif(os.name == "nt", reason="chmod is a no-op on Windows")
     @pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


Enough1122:

**Overall:** Clean, minimal test-only change making two doctor journal assertions portable across OSes (path separators and error text differ on Windows vs POSIX).

**What it does**
- `test_missing_file_keeps_the_os_error_text` now asserts `str(db) in reason` instead of the POSIX-specific "No such file or directory" text.
- `test_lists_every_managed_database` builds the nested path with `os.path.join` so the assertion matches on both separators.
- No production code changed.

**Non-blocking notes**
- The first assertion is now weaker (only checks the path appears in the reason, not that it's a "missing file" style error). If you want to keep some signal, you could additionally assert a generic error keyword (`error`, `fail`, or `No such`) — but the change is acceptable as a portability fix.
- On Windows the error text for a missing file is "The system cannot find the file specified." — the relaxed assertion correctly sidesteps it.

No functional issues found.

monerostar:

Host: Windows 11 (build 26200), Python 3.11.15, pytest 9.1.1.

This PR cannot be collected on Windows by itself. Current `main` still has a bare `os.geteuid()` skipif in the same file, so pytest dies with `AttributeError: module 'os' has no attribute 'geteuid'` before these assertions run. That collection abort is https://github.com/NousResearch/hermes-agent/pull/97801. I stacked this diff on top of #97801 to actually execute the two tests.

### `test_lists_every_managed_database` — works

On this host the report line uses `kanban\boards\myboard\kanban.db`. `os.path.join(...)` matches. That failure is gone (32 passed / 3 skipped with this hunk applied, vs 31/3 with #97801 alone).

### `test_missing_file_keeps_the_os_error_text` — still fails

`assert str(db) in reason` is still false on Windows:

```
str(db) =
  C:\Users\Admin\AppData\Local\Temp\pytest-of-Admin\pytest-178\...\gone.db

reason =
  [WinError 2] The system cannot find the file specified: 'C:\\Users\\Admin\\AppData\\Local\\Temp\\pytest-of-Admin\\pytest-178\\...\\gone.db'
```

`OSError` on Windows puts a `repr()`-style path in the message, so backslashes are doubled. `str(WindowsPath(...))` uses single backslashes. They never match.

`gone.db` itself does appear in `reason`. A portable form that passed the same live error string here:

```python
db = tmp_path / "gone.db"
reason = doctor._unreadable_reason(db)
assert db.name in reason
```

That also holds on POSIX (`No such file or directory: '.../gone.db'`). If you want to keep an "this is a missing-file error" signal without pinning one OS's wording, `db.name` plus something like `"WinError" in reason or "No such file" in reason` would do it — I would not assert the full `str(db)`.

### Takeaway

Needed together with #97801, and the nested-path hunk is good. Please don't merge the `str(db) in reason` change as-is; it does not go green on native Windows.

v1shay:
Thanks for the native Windows verification, @monerostar. I updated the missing-file assertion in e370d0a271 to check `db.name` instead of the full path, since Windows escapes path separators in the OSError text. I’ll keep this PR focused on the assertion portability fixes; the collection blocker is tracked separately in #97801. The nested database-path assertion remains covered by `os.path.join`.

alt-glitch:

Duplicate of #95122: it already covers the same doctor-test portability seams, including collection-time geteuid guarding, and has passing Windows CI.

PR #97537, opened 2026-08-29
fix(gateway): normalize background topic sources

### LINKED ISSUE #97498: Telegram DM topic recovery is skipped for /background dispatch
## Bug description

`/background` (including `/bg` and `/btw`) dispatches the raw `event.source` instead of the Telegram DM-topic recovery-normalized source. When Telegram supplies a lobby-shaped command event with an empty `thread_id`, normal foreground turns recover the user's bound topic, but the background child receives no topic identity.

This affects more than completion delivery. The child resolves session-scoped model, reasoning, service-tier, channel prompt, and toolset behavior from the wrong source/session key, and tools that require the current topic cannot act safely.

## Current source

Reproduced against `NousResearch/hermes-agent@3340bbbdad8368e7f3d9f6827d61692adbbd87d6`.

`GatewayRunner._handle_background_command()` currently does:

```python
source = event.source
```

Normal foreground handling and session-scoped commands use `_normalize_source_for_session_key()` to apply Telegram DM-topic recovery before deriving routing or session state.

## Reproduction

Add this behavioral regression to `tests/gateway/test_background_command.py`:

```python
@pytest.mark.asyncio
async def test_telegram_topic_source_is_normalized_before_dispatch():
    runner = _make_runner()
    event = _make_event(text="/background rename this topic")
    recovered = SessionSource(
        platform=Platform.TELEGRAM,
        user_id="12345",
        chat_id="67890",
        user_name="testuser",
        thread_id="42",
    )
    runner._normalize_source_for_session_key = MagicMock(return_value=recovered)
    runner._run_background_task = AsyncMock()

    await runner._handle_background_command(event)
    await asyncio.gather(*runner._background_tasks)

    runner._normalize_source_for_session_key.assert_called_once_with(event.source)
    assert runner._run_background_task.await_args.args[1].thread_id == "42"
```

Run:

```text
python -m pytest tests/gateway/test_background_command.py::TestHandleBackgroundCommand::test_telegram_topic_source_is_normalized_before_dispatch -q
```

Observed failure:

```text
AssertionError: Expected 'mock' to be called once. Called 0 times.
```

## Expected behavior

Before starting the child, `/background` should normalize the source using the same Telegram DM-topic recovery path as normal foreground turns. The recovered source should be used consistently for:

- Background execution
- Session-key and session-scoped model/reasoning/service-tier resolution
- Channel prompt and toolset resolution
- Completion delivery metadata

## Actual behavior

The raw source is passed to `_run_background_task()`. A stripped Telegram topic event therefore creates a background child without the active `thread_id`, even though the foreground gateway can recover it.

## Suggested fix

Normalize once at command receipt and pin that result for the entire task:

```python
source = await asyncio.to_thread(
    self._normalize_source_for_session_key,
    event.source,
)
```

The async offload matches existing session-scoped command handlers because topic recovery may read session bindings.

## Related work

- #83820 preserves structured reply and topic context for `/background`, but its current patch still initializes `source = event.source`, so this recovery gap appears to remain.
- #75789 covers the broader policy question for Telegram DM-topic recovery. This report is narrower: once that recovery policy applies, `/background` skips the established normalization path entirely.

wesleysimplicio:
Closes #97498

Normalize the source used by `/background` before scheduling the task so Telegram DM topic recovery uses the same session key as normal message turns. Adds regression coverage.

## Changes
- call `_normalize_source_for_session_key` before creating the background task
- preserve the normalized `SessionSource` (including recovered `thread_id`) for downstream scheduling
- add a regression test that asserts normalization happens once and the scheduled task receives the normalized source
- restore `gateway/slash_commands.py` to a reviewable UTF-8 text diff; the implementation change is limited to the normalization call

## Validation
- `uv run --extra dev pytest -q tests/gateway/test_background_command.py` — 7 passed
- `uv run ruff check gateway/slash_commands.py tests/gateway/test_background_command.py` — all checks passed
- `git diff --check` — passed


Enough1122:

**Overall:** Fixes background command scheduling so the session key uses the recovered Telegram topic source (normalized) rather than the raw event source. The test is solid; note the source file diff itself is opaque.

**What it does**
- `gateway/slash_commands.py` — the diff shows as **binary** ("Binary files a/... and b/... differ"), so the actual implementation change is not textually reviewable here. Presumably `_handle_background_command` now calls `_normalize_source_for_session_key(event.source)` before scheduling.
- New test `test_background_normalizes_source_before_scheduling` patches `_normalize_source_for_session_key` to return a `replace(event.source, thread_id="topic-42")` and asserts `_run_background_task` receives that normalized source (thread_id preserved) and that normalize was called exactly once with the raw source.

**Non-blocking notes**
- The binary diff (likely a line-ending/encoding rewrite of the whole file) means the change can't be reviewed from the patch — please confirm the committed `slash_commands.py` diff is limited to the normalization call and doesn't hide an accidental whole-file churn. This is the main thing to double-check.
- Test-only visibility: the patch verifies the normalize-then-schedule wiring, but not the actual normalization logic (which lives in existing code covered elsewhere). Consider whether a direct test of the topic-key recovery path adds value.
- `await asyncio.sleep(0)` to let the scheduled task run is fine; the patch on `_run_background_task` with `AsyncMock` cleanly isolates the assert on `await_args.args[1]`.

No functional concerns based on the visible test; verify the opaque source diff.

wesleysimplicio:
Confirmed and fixed the opaque diff concern. The branch now restores `gateway/slash_commands.py` as UTF-8 text and the source diff is limited to the normalization call (`+3/-1`); the test-only change remains separate and reviewable. Validation: `tests/gateway/test_background_command.py` — 7 passed; Ruff — all checks passed; `git diff --check` — passed.

wesleysimplicio:
CI caught a test-only issue in the new regression: the assertion read `await_args` after the patched method had been restored. Fixed in the latest commit by retaining the `AsyncMock` in a local variable. Local validation now passes: `tests/gateway/test_background_command.py` — 7 passed; Ruff — all checks passed; `git diff --check` — passed. The new CI run is queued.

wesleysimplicio:
The corrected commit is now green across the required remote checks: CI, Nix flake check, and Docker build/test. The earlier test failure is resolved.

PR #97539, opened 2026-08-29
fix(web): touch scrolling for dashboard chat + suppress iOS pull-to-refresh

ChronodeAi:
## Problem

On iOS (Safari/Brave), the dashboard's embedded Chat tab cannot be scrolled by touch:

- xterm.js only translates mouse-wheel events into scrollback — there is no touch scrolling.
- The app shell is a fixed-height layout (`h-dvh max-h-dvh overflow-hidden`), so the page itself never scrolls.

A downward drag therefore does nothing, and the gesture falls through to the browser's pull-to-refresh, which reloads the page. Disabling pull-to-refresh in browser settings only removes the reload — the scroll is still dead.

## Fix

**Commit 1 — scroll fix:**

1. Touch bridge in `ChatPage.tsx`, beside the existing wheel bridge (`attachCustomWheelEventHandler`): single-finger drags map to `term.scrollLines()`; `preventDefault` on `touchmove` keeps the gesture inside the terminal; two-finger gestures are ignored so pinch-zoom keeps working. Listener cleanup registered alongside the effect's other disposables. Sensitivity via module constant `TOUCH_SCROLL_PX_PER_ROW = 3` (≈2.5× a 1:1 drag — flick-friendly on phones).
2. `overscroll-behavior-y: none` on `html, body` in `index.css`, so drag attempts at the top of a page no longer fire pull-to-refresh (supported since Safari 16). Inner scroll containers (xterm pane, lists) are unaffected.

**Commit 2 — label rename:** the chat side panel holds the session list and the model picker, so "Model & tools" → "Sessions & model".

## Testing

- `tsc -b` and `vite build` pass clean.
- Manually verified against a remote `hermes dashboard` backend on an iPhone (Brave iOS): chat scrollback scrolls by drag, pull-to-refresh no longer fires from scroll attempts, pinch-zoom unaffected, desktop wheel behavior unchanged.


Enough1122:

**Overall:** Small, targeted web/mobile fix combining three changes: iOS pull-to-refresh suppression on the root scroller, touch-scroll bridging for the xterm chat pane, and an i18n label correction. Each is self-contained.

**What it does**
- `web/src/index.css` — `overscroll-behavior-y: none` on `html, body` kills pull-to-refresh/rubber-band on the dashboard root; inner scrollers are unaffected (they set their own `overscroll-contain`). Requires iOS 16+ (Safari 16) as documented; older iOS silently ignores it.
- `web/src/pages/ChatPage.tsx` — single-finger touch drags on the xterm host are bridged into `term.scrollLines(rows)` with `TOUCH_SCROLL_PX_PER_ROW = 3` (~2.5× speed, tuned for phones). `touchmove` uses `{ passive: false }` so `preventDefault()` keeps the drag from escaping to the page; two-finger gestures are left alone for browser zoom. Cleanup added to the component teardown.
- `web/src/i18n/en.ts` — swaps the model tools sheet title/subtitle ("Model & tools" → "Sessions & model").
- Tests: none for the touch logic (interaction-level), which is acceptable for this kind of change.

**Non-blocking notes**
- `preventDefault()` fires on every `touchmove` where `dy !== 0`, even when `rows` truncates to 0 — this blocks native page scroll while dragging inside the terminal even for sub-pixel moves. Intended (keeps drag inside the pane), but confirm it doesn't fight the browser when the terminal is shorter than the viewport and there's nothing to scroll.
- The touch bridge consumes single-finger drags, which on mobile typically also means text selection inside the terminal. If long-press selection is needed on touch, verify it still works (xterm's touch selection is separate from drag).
- `touchstart` is `{ passive: true }` while `touchmove` is `{ passive: false }` — correct (only `touchmove` needs preventDefault).
- The 3px/row constant is a magic number; the comment explains the tuning rationale, but consider a brief note on how it interacts with the wheel bridge's row math above.

No functional issues found. Well-commented, low-risk change.

ChronodeAi:
Closing per operator preference — keeping this in our fork rather than upstream.

PR #97541, opened 2026-08-29
fix(mcp): fully redact bearer credentials in diagnostics

### LINKED ISSUE #97460: security: MCP test and dashboard errors can expose Bearer fragments

wesleysimplicio:
Closes #97460

Fully redact MCP Authorization/Bearer values in CLI probe output and dashboard/API error payloads, while retaining the environment-variable template on disk. Adds regression coverage for auth display and probe failures.

### DIFF
--- a/hermes_cli/mcp_config.py
+++ b/hermes_cli/mcp_config.py
@@ -25,7 +25,10 @@ from hermes_cli.config import (
 )
 from hermes_cli.colors import Colors, color
 from hermes_constants import display_hermes_home
-from hermes_cli.mcp_security import validate_mcp_server_entry
+from hermes_cli.mcp_security import (
+    redact_mcp_sensitive_text,
+    validate_mcp_server_entry,
+)
 from tools.mcp_tool import _ENV_VAR_PATTERN, _env_ref_name
 
 logger = logging.getLogger(__name__)
@@ -565,7 +568,7 @@ def cmd_mcp_add(args):
     try:
         tools = _probe_single_server(name, server_config)
     except Exception as exc:
-        _error(f"Failed to connect: {exc}")
+        _error(f"Failed to connect: {redact_mcp_sensitive_text(exc)}")
         if _confirm("Save config anyway (you can test later)?", default=False):
             server_config["enabled"] = False
             if _save_mcp_server(name, server_config):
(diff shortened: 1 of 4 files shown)


alt-glitch:

Duplicate of #97466. Both address the same MCP probe and dashboard diagnostic redaction mechanism for #97460.

PR #96386, opened 2026-08-27
fix(desktop): keep busy tile transcripts stable

Dolverin:
## What does this PR do?

Hardens the workspace-tile transcript reconciliation introduced by #95600 / #94255 so a persisted snapshot cannot replace a tile's live output or make unrelated tabs jump.

`reconcileTileTranscripts()` currently gates every tile with the main pane's global `$busy` state. That produces both failure directions:

- an idle tile is not reconciled while the main pane is busy;
- a busy tile is still reconciled while the main pane is idle, so a stale persisted snapshot can overwrite its live transcript.

The async read also captures tile presence before `getLatestSessionMessages()`, allowing a result to land after the tile becomes busy, closes, or is rebound.

## Changes

- decide liveness per tile runtime from `$sessionStates` (`busy`, `awaitingResponse`, `needsInput`, or `turnLive`);
- check that runtime-specific liveness both before and after the async storage read;
- re-read the live stored-session/runtime pair after the await so closed or rebound tiles discard stale results;
- prune signature entries for tiles that have closed, including after a successful earlier reconciliation;
- preserve idle-tile refresh while the main pane is busy;
- leave the main-pane backfill and empty-transcript behavior unchanged.

This is complementary to #95770 (cheap no-change fetch gating) and #96215 (warm-cache refresh on explicit Bot Chat reopen); neither protects an active tile from an in-flight persisted snapshot.

## Regression proof

On current `origin/main` / `v0.20.6`, with only a test-harness compatibility default for the old `busyRef` signature:

```text
use-background-sync.test.ts: 5 failed, 26 passed
```

The five failures are the intended contracts:

1. idle tile still reconciles while the main pane is busy;
2. busy tile is skipped while the main pane is idle;
3. tile becoming busy during the read rejects the snapshot;
4. tile closing during the read rejects the snapshot;
5. closed-tile signature is pruned.

On this branch:

```text
use-background-sync.test.ts: 31 passed
TypeScript typecheck (all three desktop tsconfigs): passed
Scoped ESLint: passed
Production desktop build + assert-dist-built: passed
git diff --check: passed
```

## Scope

Only these files change:

- `apps/desktop/src/app/contrib/hooks/use-background-sync.ts`
- `apps/desktop/src/app/contrib/hooks/use-background-sync.test.ts`

PR #97542, opened 2026-08-29
fix(desktop): keep busy tile transcripts stable

Dolverin:
## Summary
- reconcile background workspace tiles against each tile runtime state instead of the global main-pane busy flag
- discard async transcript snapshots when a tile becomes live, closes, or is rebound while storage is being read
- prune stale tile transcript signatures so closed tiles cannot retain reconciliation state

## Root cause
`reconcileTileTranscripts` used the global `$busy` presentation state for every open tile even though `$sessionStates` is authoritative per runtime. This let persisted transcript refreshes overwrite a tile while its own turn was live, which could empty and rebuild message views when several session tabs were open. The async storage read also lacked a post-read presence/binding check.

## Test plan
- `npx vitest run --project ui src/app/contrib/hooks/use-background-sync.test.ts` — 31 passed
- `npm run typecheck` — passed
- `git diff --check origin/main...HEAD` — passed

Regression coverage includes an idle tile while the main pane is busy, a busy tile while the main pane is idle, tiles becoming busy during a read, tiles closing during a read, and stale signature cleanup.

### DIFF
--- a/apps/desktop/src/app/contrib/hooks/use-background-sync.test.ts
+++ b/apps/desktop/src/app/contrib/hooks/use-background-sync.test.ts
@@ -14,9 +14,11 @@ import {
 } from '@/store/session'
 import {
   $attentionSessionIds,
+  $sessionTiles,
   $stalledSessionIds,
   $workingSessionIds,
   clearAllSessionStates,
+  publishSessionState,
   SESSION_WATCHDOG_TIMEOUT_MS
 } from '@/store/session-states'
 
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #96386. Both patches implement the same per-tile busy-state and stale async-snapshot guards for Desktop transcript reconciliation.

Dolverin:
Closing as superseded by merged #101214, which incorporates the transcript-refresh fix carried by #96386. The fork branch is retained for reference.

PR #97543, opened 2026-08-29
fix(kanban): coerce legacy timestamp values on read

### LINKED ISSUE #97455: [Bug]: Non-epoch timestamps written into INTEGER columns make `kanban show` crash with `invalid literal for int()

wesleysimplicio:
Closes #97455

Normalize epoch, numeric-string, and ISO-8601 timestamp values at Kanban read boundaries so corrupted legacy rows remain visible to CLI and Desktop callers. Adds regression coverage for task runs, comments, and events.

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -1,10324 +1,586 @@
-"""SQLite-backed Kanban board for multi-profile, multi-project collaboration.
-
-In a fresh install the board lives at ``<root>/kanban.db`` where
-``<root>`` is the **shared Hermes root** (the parent of any active
-profile). Profiles intentionally collapse onto a shared board: it IS
-the cross-profile coordination primitive. A worker spawned with
-``hermes -p <profile>`` joins the same board as the dispatcher that
-claimed the task. The same applies to ``<root>/kanban/workspaces/`` and
-``<root>/kanban/logs/``.
-
-**Multiple boards (projects):** users can create additional boards to
-separate unrelated streams of work (e.g. one per project / repo / domain).
-Each board is a directory under ``<root>/kanban/boards/<slug>/`` with
-its own ``kanban.db``, ``workspaces/``, and ``logs/``. All boards share
-the profile's Hermes home but are otherwise isolated: a worker spawned
-for a task on board ``atm10-server`` sees only that board's tasks,
-cannot enumerate other boards, and its dispatcher ticks don't touch
-other boards' DBs.
-
-The first (and for single-project users, only) board is ``default``.
-For back-compat its on-disk DB is ``<root>/kanban.db`` (not
-``boards/default/kanban.db``), so installs that predate the boards
-feature keep working with zero migration. See :func:`kanban_db_path`.
-
-Board resolution order (highest precedence first, all optional):
-
-* ``board=`` argument passed directly to :func:`connect` / :func:`init_db`
-  (explicit â€” used by the CLI ``--board`` flag and the dashboard
-  ``?board=...`` query param).
-* ``HERMES_KANBAN_BOARD`` env var (used by the dispatcher to pin workers
-  to the board their task lives on â€” workers cannot see other boards).
-* ``HERMES_KANBAN_DB`` env var (pins the DB file path directly â€” legacy
-  override still honoured; highest precedence when the file path itself
-  is what the caller wants to force).
-* ``<root>/kanban/current`` â€” a one-line text file holding the slug of
-  the "currently selected" board. Written by ``hermes kanban boards
-  switch <slug>``. When absent, the active board is ``default``.
-
-In standard installs ``<root>`` is ``~/.hermes``. In Docker / custom
-deployments where ``HERMES_HOME`` points outside ``~/.hermes`` (e.g.
-``/opt/hermes``), ``<root>`` is ``HERMES_HOME``. Legacy env-var
-overrides still work:
-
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #97462. The intended reader-tolerance repair matches the earlier focused implementation; this PR's destructive file rewrite is not reviewable as submitted.

PR #97546, opened 2026-08-29
fix(background-review): share memory maintenance policy

### LINKED ISSUE #97471: fix(background-review): apply memory maintenance policy to combined reviews
## Summary

Background memory review has two prompt paths:

- `_MEMORY_REVIEW_PROMPT` when only the memory nudge fires;
- `_COMBINED_REVIEW_PROMPT` when memory and skill review fire together.

On current `main`, both paths are save-oriented and neither requires inspection or consolidation of existing memory. PR #59826 addresses #59823 by adding an audit → classify → consolidate → save workflow to `_MEMORY_REVIEW_PROMPT`, but it leaves `_COMBINED_REVIEW_PROMPT` unchanged.

As a result, when both triggers fire, prompt selection bypasses the new maintenance workflow and the memory portion of the review remains a one-way save path.

Both paths should share one conservative memory-maintenance contract. Any consolidation must preserve every semantically unique fact, qualifier, exception, preference strength, and temporal condition. If safe equivalence cannot be established, the reviewer should leave the entries unchanged rather than compress or delete them speculatively.

## Reproduction and evidence

Prompt selection is explicit in `agent/background_review.py`:

```python
if review_memory and review_skills:
    prompt = getattr(agent, "_COMBINED_REVIEW_PROMPT", _COMBINED_REVIEW_PROMPT)
elif review_memory:
    prompt = getattr(agent, "_MEMORY_REVIEW_PROMPT", _MEMORY_REVIEW_PROMPT)
else:
    prompt = getattr(agent, "_SKILL_REVIEW_PROMPT", _SKILL_REVIEW_PROMPT)
```

On current `main`, the memory-only prompt says:

> Review the conversation above and consider saving to memory if appropriate.

The combined prompt's memory section says:

> Save facts about the user and durable preferences with the memory tool.

Neither prompt defines inspection or consolidation as a successful review outcome. PR #59826 rewrites only `_MEMORY_REVIEW_PROMPT`, and its prompt tests assert only that constant. The `review_memory=True, review_skills=True` selection seam is not covered.

This can be verified without mutating memory:

1. Build a background review with `review_memory=True` and `review_skills=True`.
2. Capture the prompt selected by `spawn_background_review_thread()`.
3. Confirm that it is `_COMBINED_REVIEW_PROMPT`.
4. Confirm that the prompt contains no audit/consolidation workflow and no lossless-merge invariant.

## Actual behavior

- On current `main`, both memory-review paths are save-only.
- With PR #59826 applied, the memory-only path receives the maintenance workflow, but the combined path still receives the old save-only memory instructions.
- Existing semantic duplicates may therefore remain untouched whenever both review triggers fire.

## Expected behavior

Memory-only and combined reviews should receive the same memory-maintenance policy. The presence of a simultaneous skill review should not change whether existing memory is inspected and conservatively consolidated.

## Root cause

Memory-review policy is duplicated across two prompt constants. A change to `_MEMORY_REVIEW_PROMPT` does not automatically reach `_COMBINED_REVIEW_PROMPT`.

The current task framing is also append-oriented: finding a new fact is the signal to act, while "Nothing to save" is the no-op condition. Store quality is not itself a review outcome, so there is no stable obligation to inspect or consolidate existing entries before capacity is exhausted.

This is not a missing core-tool capability. The background review fork shares the parent's memory store, receives the frozen memory snapshot, and can already issue an atomic `operations` batch whose final state is checked against the character budget.

## Proposed fix

1. Define one shared memory-maintenance instruction block and compose it into both `_MEMORY_REVIEW_PROMPT` and `_COMBINED_REVIEW_PROMPT`.
2. If the read-only audit action from PR #59826 is adopted, require both prompt paths to use audit → classify → consolidate → save.
3. Make the consolidation invariant explicit:
   - merge only exact duplicates, entries where one fully subsumes another, or entries that can be combined without changing meaning;
   - preserve every unique fact, qualifier, exception, preference strength, and temporal condition in the union entry;
   - use one atomic batch to replace one source with the semantic union and remove only the redundant sources;
   - do not merge contradictions or context-dependent preferences speculatively;
   - when equivalence is uncertain, leave the entries unchanged and report the uncertainty.
4. Add behavior-contract tests for both paths, including the real selection seam:
   - a memory-only review contains the maintenance contract;
   - `review_memory=True, review_skills=True` selects a combined prompt containing the same contract;
   - tests assert required relationships and behavior rather than snapshotting the entire prompt.
5. Keep the selected review prompt as a user-message suffix on the inherited history. Do not mutate the parent system prompt or toolset mid-conversation.

## Scope

No new core tool is needed. Existing atomic memory operations are sufficient for consolidation. If PR #59826's read-only audit action is accepted, this issue only requires that the same action and policy be applied consistently across both review paths.

## Related issues

- #59823 — directly tracks the one-way memory nudge and proposes audit → classify → consolidate → save.
- PR #59826 — current implementation of #59823; it updates only `_MEMORY_REVIEW_PROMPT`. Separate concerns about live-state freshness and audit-only mutation reporting are already recorded on #59823 and are outside this issue's scope.
- #30220 / PR #30227 — rewrites the same prompt constants for SKILL / USER.md / MEMORY.md routing and one-fact-one-store behavior. This is the same edit surface but a different behavior contract; the changes should be coordinated so both fixes compose.
- #93623 — broader duplicate detection, periodic consolidation, and maintenance tooling.
- #78307 — broader lifecycle management, staged consolidation, rollback, and protected-class UX.
- #60900 (#60902 duplicate) — tracks quota-recovery behavior that can shorten a pending write until it loses the user's intent. That error-guidance problem should remain separately scoped.
- #97316 — tracks `current_entries` response inflation after an aborted batch; adjacent response-size problem, not this prompt-selection gap.

## Environment

- Repository: `NousResearch/hermes-agent`
- Verified against `origin/main` at 7d1c9aeab739842fa42112c95d086a9897f2d513 (2026-08-28)
- Affected component: Agent Core (`agent/background_review.py`)

wesleysimplicio:
Closes #97471

Share a conservative audit → classify → consolidate → save memory policy between memory-only and combined background reviews. The policy preserves qualifiers and refuses uncertain merges; tests cover both prompt paths.

### DIFF
--- a/agent/background_review.py
+++ b/agent/background_review.py
@@ -462,9 +462,25 @@ def _digest_history(messages_snapshot: List[Dict], tail: int = 24) -> List[Dict]
 # the user-message that the forked review agent receives.  AIAgent exposes
 # them as class attributes (``_MEMORY_REVIEW_PROMPT`` etc.) for back-compat;
 # the actual text lives here so future edits are one-place.
+_MEMORY_MAINTENANCE_POLICY = (
+    "Memory maintenance policy: audit the existing memory entries before "
+    "deciding what to write. Classify candidates as exact duplicates, fully "
+    "subsumed entries, safe semantic unions, contradictions, or unrelated "
+    "context-dependent preferences. Consolidate only exact duplicates or "
+    "entries whose union preserves every unique fact, qualifier, exception, "
+    "preference strength, and temporal condition. Use one atomic operations "
+    "batch to replace redundant entries with the lossless union and remove "
+    "only the sources it replaces. Never merge contradictions or context-"
+    "dependent preferences speculatively; when equivalence is uncertain, "
+    "leave the entries unchanged and report the uncertainty. Then save any "
+    "new durable facts.\n\n"
+)
+
+
 _MEMORY_REVIEW_PROMPT = (
     "Review the conversation above and consider saving to memory if appropriate.\n\n"
-    "Focus on:\n"
+    + _MEMORY_MAINTENANCE_POLICY
+    + "Focus on:\n"
     "1. Has the user revealed things about themselves — their persona, desires, "
     "preferences, or personal details worth remembering?\n"
     "2. Has the user expressed expectations about how you should behave, their work "
@@ -618,7 +634,8 @@ _COMBINED_REVIEW_PROMPT = (
     "desires, preferences, personal details, or expectations about "
     "how you should behave? Save facts about the user and durable "
     "preferences with the memory tool.\n\n"
-    "**Skills**: how to do this class of task. Be ACTIVE — most "
+    + _MEMORY_MAINTENANCE_POLICY
+    + "**Skills**: how to do this class of task. Be ACTIVE — most "
     "sessions produce at least one skill update. A pass that does "
     "nothing is a missed learning opportunity, not a neutral outcome.\n\n"
     "Target shape of the skill library: CLASS-LEVEL skills with a rich "
--- a/tests/run_agent/test_review_prompt_class_first.py
+++ b/tests/run_agent/test_review_prompt_class_first.py
@@ -76,6 +76,25 @@ def test_combined_review_prompt_has_memory_section():
     assert "memory tool" in prompt
 
 
+def test_memory_maintenance_policy_is_shared_by_memory_review_paths():
+    """Combined reviews must not bypass the lossless memory workflow."""
+    required = (
+        "audit",
+        "classify",
+        "Consolidate",
+        "atomic operations batch",
+        "every unique fact",
+        "when equivalence is uncertain",
+    )
+    for label, prompt in (
+        ("memory-only", AIAgent._MEMORY_REVIEW_PROMPT),
+        ("combined", AIAgent._COMBINED_REVIEW_PROMPT),
+    ):
+        lower = prompt.lower()
+        for phrase in required:
+            assert phrase.lower() in lower, f"{label}: missing {phrase!r}"
+
+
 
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Good refactor that extracts the memory-maintenance policy into one shared constant and prepends it to both the memory-only and combined background-review prompts, so combined reviews no longer bypass the lossless consolidation workflow.

**What it does**
- `agent/background_review.py` introduces `_MEMORY_MAINTENANCE_POLICY` and concatenates it into `_MEMORY_REVIEW_PROMPT` (after the "Review the conversation above" preamble, before "Focus on:") and `_COMBINED_REVIEW_PROMPT` (before the "**Skills**" section).
- Tests assert both prompts contain the policy's key phrases (audit/classify/Consolidate/atomic operations batch/every unique fact/when equivalence is uncertain), guarding against future prompt drift.

**Non-blocking notes**
- The policy text has a minor grammar slip: "Use one atomic operations batch" should read "Use one atomic operation batch" (or "an atomic operations batch"). Cosmetic, but this string now ships in every memory review prompt.
- The policy is a plain constant concatenated at import time into two prompts — single source of truth, which is the point. Just be aware the same policy is also expected verbatim in class-first tests (`AIAgent._MEMORY_REVIEW_PROMPT` etc.), so wording changes require coordinated test updates.
- The duplication guard is welcome — combined reviews previously could silently skip the audit step.

No functional issues found. Minimal, well-targeted change.

PR #95830, opened 2026-08-26
test(update): isolate the launchd fleet-restart tests from live service state

liuhao1024:
## What does this PR do?

Fixes #95827: three tests in `tests/hermes_cli/test_update_launchd_fleet_restart.py` failed on any macOS host with a real Hermes launchd gateway loaded (`scripts/run_tests.sh` reported 24 passed / 3 failed at `68518c1f9`). All three are test-isolation gaps, fixed here without touching implementation:

1. **`test_all_profiles_returns_every_gateway_service_pid`** — the all-profiles belt-and-suspenders branch runs a REAL bare `launchctl list` prefix scan (`subprocess.run` directly, unmocked), so the host's live `ai.hermes.gateway` service PID leaked into the `{100, 200}` assertion. The `_wire` fixture now serves a canned `launchctl list` listing (asserting the subprocess call is the bare scan), making the test host-independent. A new sibling pin (`test_all_profiles_scan_merges_unmapped_agent_pids`) covers the merge behavior the live scan was accidentally exercising — the #74075 prefix scan must still add PIDs for agents the label derivation can't map, pinned with a deterministic listing (`{100, 200, 300}`).
2. **`test_launchd_labels_get_launchctl_hint`** — asserted the prior `launchctl kickstart -k` recovery contract, but the implementation deliberately emits `launchctl bootstrap` since #88848 (a deregistered job cannot be kickstarted), and the macOS branch returns early — so the test passed on Linux CI while asserting a dead contract on macOS. It now pins `gw.is_macos → True` (the warning re-imports the predicate at call time, so patching the gateway attribute works on every host) and asserts the current contract: `launchctl bootstrap` present, `launchctl kickstart` absent.
3. **`test_systemd_units_keep_systemctl_hint`** — never patched the platform predicate, so a macOS host emitted launchd recovery lines even for the systemd fixture and the `launchctl not in out` assertion failed. Now pins `gw.is_macos → False`.

## Related Issue

Fixes #95827

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

[...]

PR #97547, opened 2026-08-29
fix(update): ignore successful receipt stop reasons

tachyon-r:
Treat successful command-boundary update receipts as completed rather than stopped, using a shared receipt-reason constant across update paths. Add isolated fleet-restart coverage that does not depend on the host launchd state.

### DIFF
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -2279,7 +2279,9 @@ def cmd_update(args):
         _finalize_update_receipt(1, f"{type(_update_exc).__name__}: {_update_exc}")
         raise
     else:
-        _finalize_update_receipt(0, "completed at command boundary")
+        from hermes_cli.update_receipt import COMMAND_BOUNDARY_STOP_REASON
+
+        _finalize_update_receipt(0, COMMAND_BOUNDARY_STOP_REASON)
         _update_handoff_exit_code = 0
     finally:
         _update_lock.release()
--- a/hermes_cli/update_cmd_fleet.py
+++ b/hermes_cli/update_cmd_fleet.py
@@ -78,14 +78,27 @@ def _current_checkout_sha() -> str | None:
 
 
 def _receipt_looks_unfinished(receipt: dict) -> bool:
-    """True when *receipt* is from an update that did not finish cleanly."""
+    """True when *receipt* is from an update that did not finish cleanly.
+
+    ``stop_reason`` records *how* the command boundary closed the receipt
+    (``completed at command boundary``, ``sys.exit(0)``, even KeyboardInterrupt
+    on an otherwise successful run). A truthy stop_reason must not make a
+    successful receipt look unfinished, or the next ``hermes update`` retriggers
+    ``fleet_restart_pending`` from pre-pull plan SHAs.
+    """
+    exit_code = receipt.get("exit_code")
+    outcome = receipt.get("outcome")
+    if exit_code not in (0, None):
+        return True
+    if outcome in ("failed", "partial", "running"):
+        return True
     gateway_restart = receipt.get("gateway_restart")
-    return bool(
-        receipt.get("stop_reason")
-        or receipt.get("exit_code") not in (0, None)
-        or receipt.get("outcome") in ("failed", "partial", "running")
-        or (isinstance(gateway_restart, dict) and gateway_restart.get("incomplete"))
-    )
+    if isinstance(gateway_restart, dict) and gateway_restart.get("incomplete"):
+        return True
+    stop_reason = receipt.get("stop_reason")
+    if stop_reason and outcome != "success" and exit_code != 0:
+        return True
+    return False
 
 
 def _receipt_reports_stale_runtime(expected_sha: str | None = None) -> bool:
--- a/hermes_cli/update_receipt.py
+++ b/hermes_cli/update_receipt.py
@@ -19,6 +19,7 @@ from typing import Any, Optional
 logger = logging.getLogger(__name__)
 
 _RECEIPT_KEEP = 20  # keep the last N receipts per profile home
+COMMAND_BOUNDARY_STOP_REASON = "completed at command boundary"
 
 # ``hermes update`` is a single-threaded CLI command; a module singleton lets the 7k-line updater
 # record steps from any depth without threading a handle through every helper.
--- a/tests/hermes_cli/test_update_fleet_restart_pending.py
+++ b/tests/hermes_cli/test_update_fleet_restart_pending.py
@@ -84,6 +84,12 @@ def _patch_update_deps(monkeypatch, tmp_path, run_side_effect):
     monkeypatch.setattr(hermes_main, "_is_windows", lambda: False)
     monkeypatch.setattr(main_install_repair, "_is_windows", lambda: False)
     monkeypatch.setattr(
+        update_cmd, "_restart_macos_launchd_gateways", lambda *a, **k: None
+    )
+    monkeypatch.setattr(
+        update_cmd_fleet, "_restart_macos_launchd_gateways", lambda *a, **k: None
+    )
+    monkeypatch.setattr(
         hermes_main,
         "_get_origin_url",
         lambda *a, **k: "https://github.com/NousResearch/hermes-agent.git",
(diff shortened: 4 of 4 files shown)


Enough1122:

**Overall:** Correct, narrowly-scoped fix: a receipt whose `stop_reason` is exactly `"completed at command boundary"` is no longer treated as an unfinished/interrupted update, while every other stop reason (including `sys.exit(0)` and `KeyboardInterrupt`) still counts as unfinished. Prevents a spurious fleet-restart trigger after a normal command-boundary stop.

**What it does**
- `hermes_cli/update_cmd.py:_receipt_looks_unfinished` now special-cases the known-normal stop reason before the exit-code/outcome checks; `exit_code` and `outcome` handling is otherwise unchanged.
- Tests add: a success receipt with the command-boundary stop reason does not trigger `_pending_fleet_restart_needed()`; parametrized cases confirm `sys.exit(0)` and `KeyboardInterrupt` still count as unfinished.

**Non-blocking notes**
- The special-cased string `"completed at command boundary"` is a magic literal. If the runner that writes receipts ever rewords it, behavior silently reverts (safe direction — no false trigger — but the "ignore" intent would be lost). Consider hoisting to a module-level constant, ideally shared with the code that emits the stop reason.
- The guard only exempts when `stop_reason` exactly equals the known value; a prefix/suffix variation (e.g. trailing period) would not match. Tests pin the exact form, so this is consistent by contract.
- Interaction with `outcome` is sensible: even a command-boundary stop with `outcome: "failed"` still trips the unfinished check (the `outcome` check runs regardless) — correct behavior.

No functional issues found. Minimal and well-tested.

freqyfreqy:
Independent reproduction on macOS confirms this exact chain. Install: git, v0.21.0 (`63279301`), multiplexed fleet, gateway started manually (no launchd supervision).

**Shape observed:** `logs/update_receipts/latest.json` from a successful Aug 30 update (v0.20.6 → stayed current) carried `outcome=success`, `exit_code=0`, `stop_reason="completed at command boundary"` (written by the command-boundary finalizer because the already-up-to-date path exits before an inner finalize), `fleet: []`, and `plan.runtimes[0].code_sha=4f225435` (correctly captured pre-pull, per `_receipt_reports_stale_runtime`'s own docstring). After the repo advanced to `63279301`, every subsequent `hermes update` (even no-op ones) printed "A previous `hermes update` pulled new code but did not restart running gateways", ran the catch-up against zero running gateways, printed "No running gateways — nothing to restart", and **still re-fired next run** — because the catch-up only clears the marker file, which was never written; the trigger is the receipt, and it never gets rewritten. Permanent loop.

**Collateral, worth noting for the restart-via-manual case:** one of these phantom catch-ups did hit a *running* gateway (`supervisor: manual`, from the fleet snapshot). The updater stopped it expecting launchd to revive it ("⚠ launchd did not revive the gateway after its graceful exit — forcing restart"), got nothing, and the update exited 1 with "Update incomplete". So the false positive doesn't just log noise — on a manual-supervisor setup it can **take down a live gateway the pull didn't even need to restart**.

**Fix verification:** edited the receipt per this PR's semantics (treat the boundary stop reason as completed) → `_pending_fleet_restart_needed()` returns False, and a later no-op update completed cleanly with no warning and wrote a fresh, clean `latest.json`. Loop dead. The `COMMAND_BOUNDARY_STOP_REASON` constant approach is the right one — the alternative (rewriting/deleting receipts) destroys the audit log.

Local suite: `scripts/run_tests.sh tests/hermes_cli/test_update_fleet_restart_pending.py` → 13 passed, 2 failed (`test_marker_written_after_pull_cleared_after_successful_restart`, `test_clean_update_warns_about_surviving_pre_update_serve_runtime`) — both are the known live-state leakage in this suite (#98037 / #95830 territory: the captured logs show the test enumerating *real* running `serve` PIDs after the module purge), reproduced identically on unmodified `main`, so not caused by this change.

freqyfreqy:
Correction — one chronology claim above was wrong and I'd rather fix it than let it stand:

I wrote that I edited the receipt and the next no-op update then completed cleanly. Checking mtimes: the clean no-op run (21:44) **preceded** my edit by ~6 minutes, and the edit was itself a no-op — that run had already replaced `latest.json`. So the loop stopping is not attributable to my intervention; it self-discharged mechanically: that run's receipt captured no gateway runtime at all (all four recorded runtimes are `serve`s with `code_sha=null`), so `_sha_mismatch` never fires and `_receipt_reports_stale_runtime` → False regardless of the boundary stop reason. (Which also illustrates the latent gap: the receipt path only discharges when some later capture happens to record non-skewing SHAs — one skewed gateway runtime keeps it re-firing forever.)

What stands as verified:
- The bug chain reproduced live on macOS this week: three back-to-back update runs all fired the phantom catch-up on an already-current checkout, and one of them force-killed a *running* manual-supervisor gateway (updater expected launchd to revive it, got nothing, exited 1 "Update incomplete").
- Predicate-level A/B against the archived smoking-gun receipt (`stop_reason="completed at command boundary"`, `outcome=success`, `exit_code=0`, `fleet=[]`, pre-pull runtime `code_sha` ≠ current HEAD): shipped `_receipt_looks_unfinished` → True and `_receipt_reports_stale_runtime(expected_sha=HEAD)` → True (re-fires forever); with this PR's exact patched predicate → False. Independent confirmation of the fix semantics.
- The failure mode stays latent for fleets that don't let the updater restart their gateway (manual supervisor): the trigger re-arms on the next real pull, since a fresh receipt again captures pre-pull runtime SHAs and writes the boundary stop reason via the command-boundary finalizer.

On the 2 test failures I reported: those are on *unmodified* main, not this branch — likely the live-service state leakage the suite is known for (#98037 family), and your fixture change monkeypatching `_restart_macos_launchd_gateways` may already address one of them. Disregard that as a judgement on this branch.

Still +1 from me on the fix.

chenlichao:
Independent validation of this fix on top of current main (f58fcc8118):

- Root-caused the full #98022 loop live on a Linux v0.21.0 git install (systemd gateway). The purge half is #103478 (receipts now persist); this stop_reason half is what keeps re-arming the catch-up: a clean success receipt stays classified as "unfinished" forever, so `_receipt_reports_stale_runtime()` compares its pre-pull plan SHAs against every advancing HEAD.
- Ran the targeted suites on your rebased branch locally: `test_update_fleet_restart_pending.py` + `test_update_receipt.py` + `test_update_handoff_exit.py` → 48 passed. `test_successful_command_boundary_receipt_without_fleet_does_not_retrigger` is red on unfixed main, green with the fix.
- Live verification on the affected box (both halves cherry-picked onto the running install): `hermes update` writes a fresh success receipt every run, `_pending_fleet_restart_needed()` returns False, and the startup warning is gone.

One observation from the live reproduction for reviewers: `test_successful_non_boundary_stop_reasons_are_finished` treats a `KeyboardInterrupt` stop_reason on a success/exit-0 receipt as finished. The genuinely-interrupted case that motivated #95294 (Ctrl-C mid-update) records `outcome: failed` / non-zero exit (and the `fleet_restart_pending` marker when interrupted after the pull), so it still re-triggers through the other clauses and the marker — confirmed on the affected box, where an interrupted update's receipt (`stop_reason: "KeyboardInterrupt: "`, `outcome: failed`) keeps the catch-up armed with this fix in place.

kshitijk4poor:
Merged via #104153 (rebase) — your three commits are on `main` with your authorship intact: `cd27df3c2e` (fix), `6e0fba4e58` (shared `COMMAND_BOUNDARY_STOP_REASON`), `f7e56d4256` (launchd-isolated tests). On top: the predicate now names its invariant (`succeeded = exit_code == 0 or outcome == "success"`, same truth table), the literal-dict tests collapsed into one parametrized contract that also pins the `update_contract` refused-receipt shape (the one production case where the stop_reason clause is load-bearing), and `test_update_handoff_exit.py` binds to the constant.

Verified before merge with the real `_receipt_looks_unfinished` across seven receipt shapes: only a clean success and `sys.exit(0)` flip to finished; interrupt/failed/running/partial/non-zero-exit stay unfinished. Thank you @tachyon-r — and @freqyfreqy and @chenlichao for the independent reproductions on macOS and Linux that pinned this as the second half of #98022 (the purge half landed in #103478).

PR #97549, opened 2026-08-29
fix(bot-chat): refresh warm prompt capability epoch

### LINKED ISSUE #97480: [Bug]: Bot Chat capability-epoch check is unreachable for a warm agent object
`stored_prompt_capability_stale()` / `stored_bot_chat_prompt_needs_upgrade()` (`tools/bot_mode_probe.py`) are documented to catch skills/toolset/MCP/SOUL.md/roster drift "on the next message" for an eternal Bot Chat session. Both are only reachable from `_restore_or_build_system_prompt()` (`agent/conversation_loop.py:919`), which itself is only called when `agent._cached_system_prompt is None` (`agent/turn_context.py:821`) — i.e. only on a cold agent object.

The gateway and Desktop's `tui_gateway` both cache the live `AIAgent` across turns for the life of a session/connection (`_agent_cache` in `gateway/run.py`; `_sessions` in `tui_gateway/server.py`), and eviction is gated on the transport being dead. For a Desktop Bot Chat with an open window, that's never — so this branch is unreachable indefinitely, not just until "the next message."

**Repro:** patched a temporary debug log into the branch, sent two turns into an already-warm Bot Chat session, confirmed zero rebuild log lines and an unchanged `sessions.system_prompt_hash` across both. Full trace in [#72253's thread](https://github.com/NousResearch/hermes-agent/pull/72253#issuecomment-5456225900).

Note this is broader than SOUL.md: `capability_fingerprint()`'s `surface` dict also covers disabled skills, enabled toolsets, MCP config, installed skill names, bot roster, roster roles, peers, and remote roster — none of which reach a warm session either. #72253 fixes SOUL.md specifically via the compaction boundary and explicitly leaves this block untouched.

**Secondary finding, same investigation:** `capability_fingerprint()`'s `surface` dict doesn't hash MEMORY.md or USER.md content at all — those already get a working refresh at the next compaction via a separate, pre-existing check (`_cached_prompt_reflects_builtin_memory`), so they're not broken, but they'd never get the "next message" treatment even if this issue is fixed. Flagging in case it's worth folding in.

**Two possible directions, no strong opinion on which:**
1. Run the fingerprint/staleness check every turn regardless of `_cached_system_prompt` state — it's cheap (file reads + hashing), not a full rebuild — and only pay the prefix-cache-breaking rebuild when something actually changed. Matches what the existing comments describe as intended.
2. Fold Bot Mode capability staleness into the same compaction-boundary gate #72253 uses for identity, consistent with `AGENTS.md:19-23`'s "compaction is the only mid-conversation rewrite exception" rule — but that downgrades "on next message" to "at next compaction" for the whole capability surface, a narrower guarantee than currently documented.

Relates to #68563 and #72253.

wesleysimplicio:
Closes #97480\n\nChecks the Bot Chat capability epoch on every warm turn, while preserving the prompt cache when the capability surface is unchanged. Rebuilds and persists the prompt only after detected capability drift.

### DIFF
--- a/agent/conversation_loop.py
+++ b/agent/conversation_loop.py
@@ -925,75 +925,8 @@ def _restore_or_build_system_prompt(agent, system_message, conversation_history)
         # exception applied to capabilities. Prompts without the stamp
         # (every non-Bot-Chat session) never take this branch, and the check
         # fails closed to "reuse" so a probe failure can't burn cache.
-        _bot_stale = False
-        try:
-            from tools.bot_mode_probe import (
-                BOT_CHAT_TITLE,
-                stored_bot_chat_prompt_needs_upgrade,
-                stored_prompt_capability_stale,
-            )
-
-            _home_for_epoch = None
-            try:
-                from agent.system_prompt import _agent_home
-
-                _home_for_epoch = _agent_home(agent)
-            except Exception:
-                pass
-            _bot_stale = stored_prompt_capability_stale(stored_prompt, _home_for_epoch)
-            if not _bot_stale and getattr(agent, "_bot_mode_protocol", True):
-                # Legacy upgrade: a Bot Chat whose prompt predates the epoch
-                # mechanism (no stamp, no protocol) gets ONE migration
-                # rebuild — otherwise pre-existing bots would never learn
-                # the messaging protocol. Title-gated so ordinary unstamped
-                # sessions (i.e. all of them) never take this path; the
-                # rebuilt prompt carries the stamp, so it cannot re-fire.
-                _t = str(getattr(agent, "_session_title_hint", "") or "").strip()
-                if not _t and agent._session_db and agent.session_id:
-                    try:
-                        _t = str(agent._session_db.get_session_title(agent.session_id) or "").strip()
-                    except Exception:
-                        _t = ""
-                if _t == BOT_CHAT_TITLE:
-                    _bot_stale = stored_bot_chat_prompt_needs_upgrade(stored_prompt, _home_for_epoch)
-        except Exception:
-            _bot_stale = False
-        if _bot_stale:
-            logger.info(
-                "Bot Chat capability epoch changed for session %s; rebuilding "
-                "system prompt to adopt the new capability surface (one-time "
-                "prefix-cache break).",
-                agent.session_id,
-            )
-            agent._session_title_hint = "Bot Chat"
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Clean refactor plus a behavioral extension: the Bot Chat capability-epoch refresh logic is extracted into two helpers, and — the key addition — `build_turn_context` now runs the epoch check even when the system prompt is already warm-cached, so gateway/desktop agents pick up user-initiated skill/toolset/MCP changes on the next message instead of only at prompt build.

**What it does**
- `agent/conversation_loop.py` extracts the inline refresh block into `_bot_chat_prompt_needs_refresh(agent, prompt)` (probe logic; fails closed to "no refresh" on any probe error so the prefix cache is preserved) and `_refresh_bot_chat_prompt(agent, system_message)` (rebuild + persist + skills-cache clear). The restore path behavior is byte-for-byte equivalent.
- `agent/turn_context.py:build_turn_context` gains `refresh_warm_bot_chat_prompt`; when the cached prompt is non-None and the callback is provided, it runs the epoch check each turn.
- `run_conversation` wires the callback via a lambda that checks `_bot_chat_prompt_needs_refresh` against `agent._cached_system_prompt`.
- Test `test_warm_prompt_runs_bot_capability_refresh_check` asserts the refresh callback is invoked once for a warm agent.

**Non-blocking notes**
- The refresh callback runs on **every** turn for warm agents (any session type, not just Bot Chat). For non-Bot-Chat sessions `_bot_chat_prompt_needs_refresh` should return False cheaply (stamp check + `_bot_mode_protocol` guard), but confirm `stored_prompt_capability_stale` is genuinely cheap for all warm prompts — otherwise this adds per-turn overhead to every warm session.
- The lambda passes `getattr(a, "_cached_system_prompt", "")` as the prompt, while `_refresh_bot_chat_prompt` rebuilds from the real `system_message`. Consistent, but note that if the cached prompt were ever an empty string the stale-check would run against `""` — the probe should handle that gracefully (fails closed), which the extraction preserves.
- `_bot_chat_prompt_needs_refresh` keeps the nested imports and `_agent_home` tolerance from the original — fine, just verbose; no change needed.
- The refactor correctly preserves the one-time `_bot_capability_refreshed` flag and the DB persist with re-fire-on-failure semantics.

No functional issues found. The refactor is behavior-preserving and the warm-refresh addition is well-scoped.

PR #83400, opened 2026-08-10
feat(gateway): allow selective multiplex profile serving

tmchow:
Closes #83377

## Why this matters

Hermes profiles can represent different capability and trust levels, not just different personalities.

A common example is one shared gateway with:

- `default`: the normal general-purpose profile
- `kids-safe`: a restricted profile with a limited toolset, isolated configuration, and no inherited general-purpose credentials
- other developer or worker profiles that exist on the same host but should not run in this gateway

Selected conversations are explicitly routed to `kids-safe`; ordinary traffic continues to use `default`.

Before this change, `gateway.multiplex_profiles: true` made the gateway activate **every** valid named profile it found. More importantly, if traffic matched a route to `kids-safe` but that profile was unavailable, routing could silently fall back to `default`. For a restricted profile, that is not graceful degradation: it crosses the intended capability boundary.

This PR adds the missing control: explicitly choose which named profiles a multiplex gateway serves, and fail closed when a matched route targets anything outside that served set.

## Restricted-profile example

Default profile gateway configuration:

```yaml
gateway:
  multiplex_profiles: true
  multiplex_profile_allowlist:
    - kids-safe

  profile_routes:
    - name: kids-safe-messages
      platform: discord
      guild_id: "guild-demo"
      chat_id: "channel-demo"
      profile: kids-safe
```

Restricted profile hardening remains configured in that profile itself, for example:

```yaml
security:
  inherit_global_auth: false

tools:
  enabled_toolsets:
    - search
    - vision
    - tts
```

The allowlist and route solve different parts of the problem:

- `multiplex_profile_allowlist` limits this gateway to `default + kids-safe`; unrelated installed profiles are not started or exposed.
- `profile_routes` identifies which conversations belong in `kids-safe`.
- `security.inherit_global_auth: false` prevents the restricted profile from inheriting default-profile credentials.
- If the `kids-safe` route matches but that profile is missing, invalid, or not served, the message is rejected instead of running under `default`.
- If no route matches, historical default-profile routing is preserved.

This makes a shared gateway suitable for a restricted-profile deployment without pretending that profile routing alone is a complete security sandbox. Separate processes or OS isolation remain appropriate when a hard operating-system boundary is required.

## What changed

- add optional `gateway.multiplex_profile_allowlist`
- preserve historical `default + all named profiles` behavior when the setting is unset
- treat an empty list as `default` only
- normalize and deduplicate selected names
- skip invalid or missing names with warnings
- fail safely to `default` only for malformed non-list configuration
[...]

PR #97550, opened 2026-08-29
fix(cron): use gateway config for satellite delivery preflight

### LINKED ISSUE #97476: Cron preflight false-blocks Telegram delivery for satellite profiles using gateway.profile_routes

wesleysimplicio:
Closes #97476\n\nUse the multiplexing gateway's default-root delivery configuration when a cron job runs from a satellite profile, so profile-routed Telegram delivery is not falsely blocked before dispatch.

### DIFF
--- a/cron/scheduler.py
+++ b/cron/scheduler.py
@@ -5146,6 +5146,37 @@ def _preflight_check_provider_key(job: dict, cfg: dict) -> Optional[str]:
     return None
 
 
+def _load_delivery_gateway_config():
+    """Load gateway delivery config from the owning gateway process.
+
+    A multiplexed satellite profile owns its cron store but not the gateway
+    credentials.  Its local ``config.yaml`` intentionally omits credentials
+    when delivery is routed through ``gateway.profile_routes`` on the default
+    gateway.  In that topology, preflight must inspect the default root's
+    gateway config, just as fire-time routing does.
+    """
+    from gateway.config import load_gateway_config
+
+    active_home = _get_hermes_home().resolve()
+    try:
+        from hermes_constants import (
+            get_default_hermes_root,
+            reset_hermes_home_override,
+            set_hermes_home_override,
+        )
+
+        default_root = Path(get_default_hermes_root()).resolve()
+        active_home.relative_to(default_root / "profiles")
+    except (ImportError, ValueError, OSError):
+        return load_gateway_config()
+
+    token = set_hermes_home_override(str(default_root))
+    try:
+        return load_gateway_config()
+    finally:
+        reset_hermes_home_override(token)
+
+
 def _preflight_check_delivery(job: dict) -> Optional[str]:
     """Check the job's delivery target(s) resolve to configured platforms.
 
@@ -5183,9 +5214,7 @@ def _preflight_check_delivery(job: dict) -> Optional[str]:
             )
         if connected is None:
             try:
-                from gateway.config import load_gateway_config
-
-                gateway_config = load_gateway_config()
+                gateway_config = _load_delivery_gateway_config()
                 connected = {
                     p.value for p in gateway_config.get_connected_platforms()
                 }
--- a/tests/cron/test_cron_relay_delivery_guards.py
+++ b/tests/cron/test_cron_relay_delivery_guards.py
@@ -108,6 +108,24 @@ def _gateway_config(connected_values):
 
 
 class TestPreflightRelayFronted:
+    def test_satellite_profile_uses_default_gateway_delivery_config(self, monkeypatch, tmp_path):
+        """Satellite cron jobs use credentials from the multiplexing gateway."""
+        default_root = tmp_path / "hermes"
+        satellite = default_root / "profiles" / "grant"
+        satellite.mkdir(parents=True)
+        monkeypatch.setattr(sched, "_hermes_home", satellite)
+        monkeypatch.setattr(
+            "hermes_constants.get_default_hermes_root",
+            lambda: default_root,
+        )
+        with patch(
+            "gateway.config.load_gateway_config",
+            return_value=_gateway_config({"telegram"}),
+        ):
+            assert _preflight_check_delivery(
+                {"deliver": "telegram:-1004306455751:14"}
+            ) is None
+
     def test_relay_fronted_slack_accepted(self, monkeypatch):
         """Relay-only topology fronting slack: slack:CHAT passes preflight."""
         monkeypatch.setenv("GATEWAY_RELAY_PLATFORMS", "slack")


andrexibiza (commented):
Reviewed exact head `114f3f811d9a623a95699dc7bb533be609bfa8f9` against live `main@ac6c8028e00d01ee2f299ba7fd03329c7f10382d`. This is a clean 1-commit / 2-file branch, 1 ahead / 0 behind current main, with no earlier review submissions or PR-thread comments when I started.

The bug in #97476 is real, and the instinct here is right: preflight must reason about the transport that will actually deliver a satellite-profile cron result rather than blindly demanding credentials inside the job-owning profile. The current implementation crosses that boundary too broadly, though.

**P1 — `_load_delivery_gateway_config()` equates “named profile path” with “this job is owned by the default multiplex gateway.” Those are not the same authority.**

The new helper does this:

```python
active_home = _get_hermes_home().resolve()
default_root = Path(get_default_hermes_root()).resolve()
active_home.relative_to(default_root / "profiles")
...
set_hermes_home_override(default_root)
return load_gateway_config()
```

So *every* `HERMES_HOME=<root>/profiles/<name>` cron execution discards that profile's gateway configuration for delivery preflight and substitutes the default root's config. There is no check for `gateway.multiplex_profiles`, no `profile_routes` match, no served-profile check, and no proof that the default gateway is the process/transport owner.

That regresses the other legitimate topology: a named profile running independently with its own gateway/credentials. The profile architecture explicitly preserves that shape — merged #83400 says profiles excluded from one multiplexer may run standalone elsewhere — and `cron/scheduler.py::_get_hermes_home()` itself documents cron as per-profile state whose jobs are stored and executed under the active profile home. In that topology the local profile config is authoritative. With this patch, a standalone `grant` profile with its own Telegram token can be falsely blocked because default has no Telegram, or falsely accepted because default has a credential unrelated to `grant`.

The proof required here is not “path is under `profiles/`”; it is “this exact delivery is owned by the primary multiplex gateway.” Preserve the existing own-home connectivity result first, and only rescue a missing local connection when primary configuration proves the multiplex relationship. That proof needs to include the routed profile and the delivery scope, and it must respect the served-profile set established by #83400 rather than treating ambient primary credentials as authority.

**P1 — the added regression is vacuous on the defect it claims to pin.**

`test_satellite_profile_uses_default_gateway_delivery_config` patches `gateway.config.load_gateway_config` to unconditionally return a config whose connected set contains `telegram`. Current `main` already calls that exact function directly from `_preflight_check_delivery`, so the test returns `None` on unfixed main as well. It never observes which Hermes home was active when the config was loaded, and therefore never demonstrates RED → GREEN for #97476.

Please make the test model the two scopes: when `get_hermes_home()` is the satellite, `load_gateway_config()` should report Telegram disconnected; when the proven primary owner is consulted, it should expose the route/transport that can actually deliver. Then add the inverse control above: a standalone named-profile cron with local Telegram connected and default Telegram absent must remain accepted. I would also pin an unrouted/foreign-profile target so a primary Telegram credential alone cannot rescue the wrong profile.

**There is already an earlier, same-issue implementation and these branches collide.**

Open #97484 (`liuhao1024`), created before this PR, also fixes #97476 and modifies the same `cron/scheduler.py` seam. Its shape is materially narrower: it leaves the profile's own gateway config authoritative and adds a rescue only after the local platform is unconnected, consulting primary `profile_routes` and requiring the route to point at the profile currently being served. #97550 is therefore duplicate work at the issue/FILE-LIST level, not an independent feature lane. Please consolidate rather than land both competing authorities. Preserve Wesley Simplicio's contribution/credit for any useful test or simplification that survives; preserve `liuhao1024`'s earlier implementation authority on #97484 rather than silently superseding it.

There is one additional landing-boundary issue whichever branch survives: `cron/scheduler.py` is already a tracked 2K-law godfile under open #79918 / #78647, and this patch adds another reusable configuration/ownership seam directly into it. The delivery-preflight policy should land through that existing fracture graph rather than becoming more permanent scheduler ownership.

**Exact-object gates:** the sole commit is `Wesley Simplicio <wesleysimplicio@live.com>`, but current main has no contributor mapping for that email, so add the normal mapping without rewriting authorship. At review time exact-head CI `33224771427`, Docker `33224770985`, and Nix `33224770992` are still in progress; this commit is not green yet. All three need to complete successfully on the surviving exact commit before landing.

The underlying report is useful and worth closing. The key is to make preflight follow *qualified delivery ownership*, not filesystem ancestry: local profile first, exact multiplex-route rescue second, unrelated primary credentials never ambient authority. That gets #97476 fixed without breaking standalone profiles or creating a second competing implementation. 🚀

alt-glitch:

Related: #97484 fixes #97476 through a narrower primary profile_routes check; this PR instead loads the default gateway configuration. These are competing mechanisms for the same preflight failure.

teknium1:
Closing in favor of #97484, merged via #99375 (merged, commit 9eb832aad7) — same preflight false-block, but #97484 was submitted ~2h earlier and reads only the primary's `profile_routes` (targeted, fail-closed) rather than loading the primary gateway config wholesale into the satellite's preflight. Your diagnosis of the false-block was correct. Thanks!

Issue #89591, opened 2026-08-19
[Bug]: Bundled Bot Mode absent from Windows Desktop plugin inventory with remote gateway


wliu-dev:
## Bug description

The bundled **Bot Mode** plugin is absent from the Windows Desktop plugin inventory when the Desktop client is connected to a remote Hermes gateway.

This is not a disabled-plugin case: **Settings → Plugins** lists only `Example Plugin` and `Kanban`; there is no `Bots` entry to enable. The `BOTS` tab is also absent beside `SESSIONS`.

## Reproduction

1. Run a current Hermes gateway on a remote Linux host.
2. Connect current Hermes Desktop on Windows to that remote gateway.
3. Apply the Desktop update, fully quit Desktop, then force a local renderer rebuild:
   ```powershell
   hermes desktop --force-build
   ```
4. Launch the newly rebuilt `Hermes.exe`.
5. Open **Settings → Plugins**.

## Expected behavior

Per the Bot Mode documentation, `Bots` is a bundled, default-on Desktop plugin and should appear next to `Sessions`.

## Actual behavior

`Bots` is absent from both the plugin inventory and the sidebar, despite a successful current-source renderer rebuild.

## Verified evidence

- At the time of testing, Desktop and remote backend both displayed `v0.20.4 (+4)`.
- The Windows rebuild wrote the Desktop install stamp from current `main`:
  ```text
  [hermes] install stamp: c69b6471e677 (main) from local
  ```
- The freshly built renderer bundle contains the Bot Mode code:
  ```text
  PASS: Bot Mode is present in the freshly built renderer bundle.
  ```
- Therefore this is not explained by an old Electron bundle or a missing source file; the bundled plugin is apparently not being discovered/registered at runtime.

## Related issue

Related to #88870 (Bot Mode panel disappears on a remote gateway), but distinct: #88870 reports that Bots remains visible/enabled in Plugins and then the panel disappears. In this case the `Bots` plugin is absent from the Desktop plugin inventory entirely.

## Environment

- Desktop client: Windows
- Gateway: remote Linux host
- Connection shape: Windows Desktop → remote gateway

jackulau:
I went after this on Windows 11 and can rule out the two things I would have checked first. No fix yet - I do not have a root cause, and I would rather not ship one without it - but this should save whoever takes it a couple of hours, and there is one concrete change at the end that would have made your report answerable in one line.

## Ruled out: the bundled-plugin glob

[...]

PR #97551, opened 2026-08-29
fix(desktop): unwrap Rolldown CJS-interop wrapper in discoverBundledPlugins

gridnodeinfra-network:
## Summary

`discoverBundledPlugins()` reads `mod.default` from each `import.meta.glob` entry expecting a `HermesPlugin` object. Under Rolldown, that field is a **getter function** `{ default: () => HermesPlugin }` (the rolldown-runtime `u(e, t, n)` helper preserves the `default` key when the wrapper has no other source props to copy). The validation `!plugin?.id || typeof plugin.register !== 'function'` then sees a function reference (no `.id`, no `.register`) and silently skips every bundled plugin with `[plugins] X has no valid default HermesPlugin export — skipped`. Net result: `$pluginRecords` stays empty and Settings → Plugins → Desktop Plugins renders "No desktop plugins installed yet." on Windows packaged builds.

## Scope of this fix

This change only touches desktop bundled-plugin unwrapping (`apps/desktop/src/contrib/plugins.ts`). It does NOT touch:

- the backend `plugins.manage` RPC handler (`tui_gateway/methods_tools.py:2395`) — verified healthy on RackNerd, returns the expected 55-row payload
- the bottom Agent Plugins section's `isDesktopRelevantPlugin` filter — that filter intentionally returns `false` for any plugin with `source === 'bundled'`, which is why the bottom section correctly shows 0 even when 55 backend plugins are returned. This is curation behavior, not a bug.
- the gateway, OAuth, or any RPC layer

## Fix

Extract a small `unwrapPluginDefault(mod)` helper that calls the default export when it's a function and otherwise returns it directly. The `typeof === 'function'` test terminates unwrapping exactly when the helper sees the plugin object, so the helper is invariant to wrap depth and works for both the Rolldown-wrapped shape and a normal ESM direct export. The existing validation block is preserved immediately after the unwrap.

```ts
const plugin = unwrapPluginDefault(mod)

if (!plugin?.id || typeof plugin.register !== 'function') {
  console.warn(`[plugins] ${path} has no valid default HermesPlugin export — skipped`)
  continue
}
```

The `import.meta.glob<>` generic is widened to `{ default: HermesPlugin | (() => HermesPlugin) }` to accept either shape.

## Tests

- Added `apps/desktop/src/contrib/plugins.test.ts` covering: ESM shape, Rolldown shape, empty `{}` / `null` / `undefined` input, throwing getter, getter returning `null` / `undefined`.
- Standalone semantic verifier: the extracted `unwrapPluginDefault` was run against 6 cases including a regression test that replicates the original code path and confirms (a) `mod.default` fails validation for `{ default: () => realPlugin }`, (b) `unwrapPluginDefault(mod)` returns the real plugin and passes validation. All 6 pass.

## Verification status

- **Typecheck** (`tsc -p apps/desktop --noEmit`): clean for the two changed files.
- **Runtime semantics**: 6/6 standalone test cases pass.
- **Full desktop Vitest**: **not run on contributor's host** because the `apps/desktop` workspace's `node_modules` is partially absent (`driver.js`, `@tanstack/react-query`, `@testing-library/react`, desktop's own vitest binary are missing). Running `npm install --workspace apps/desktop` is heavy and OOM-prone; deferred to upstream CI.
- **Upstream CI is the authoritative test runner here.** Treat the new test as added but unverified locally; full CI must pass before merge.

## Files

| File | Change |
|------|--------|
| `apps/desktop/src/contrib/plugins.ts` | +25 / -2: widen `import.meta.glob` generic; add `unwrapPluginDefault`; use it at the existing call site |
| `apps/desktop/src/contrib/plugins.test.ts` | +45: new vitest covering both shapes and edge cases |

## Reproduction (Windows Desktop packaged build)

1. Install v0.20.6 (or v0.20.5, tag v2026.8.19) from the official installer.
2. Open Settings → Plugins. Header reads "0 installed".
3. Cmd-K → DevTools → console filter `[plugins]`: three `has no valid default HermesPlugin export — skipped` warnings.
4. After this fix is applied and a new build is installed, the same `$pluginRecords` contains three entries (accent, hermes-bots, kanban) and the section header reads "3 installed".

Closes the bundled-plugin half of #89591 / #94726 as observed on Windows packaged builds (the remote-gateway half on Linux/macOS source builds is a separate defect).

### DIFF
--- a/apps/desktop/src/contrib/plugins.test.ts
+++ b/apps/desktop/src/contrib/plugins.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, it, vi } from 'vitest'
+
+import { unwrapPluginDefault } from './plugins'
+
+vi.mock('./runtime-loader', () => ({ watchRuntimePlugins: vi.fn() }))
+vi.mock('./plugins-store', () => ({
+  publishPlugin: vi.fn(),
+  pluginActive: vi.fn().mockReturnValue(false)
+}))
+
+describe('unwrapPluginDefault', () => {
+  it('returns the plugin object when default is the ESM object form', () => {
+    const plugin = { id: 'accent', name: 'Accent', register: vi.fn() }
+    // Normal Vite/ESM glob result: `{ default: HermesPlugin }`
+    expect(unwrapPluginDefault({ default: plugin })).toBe(plugin)
+  })
+
+  it('invokes the default getter and returns the plugin when it is a Rolldown CJS-interop wrapper', () => {
+    const plugin = { id: 'hermes-bots', name: 'Bots', register: vi.fn() }
+    const getter = vi.fn(() => plugin)
+    // Rolldown CJS-interop: `{ default: () => HermesPlugin }`
+    const result = unwrapPluginDefault({ default: getter })
+
+    expect(getter).toHaveBeenCalledTimes(1)
+    expect(result).toBe(plugin)
+  })
+
+  it('returns null when the module shape has no default export', () => {
+    expect(unwrapPluginDefault({})).toBeNull()
+    expect(unwrapPluginDefault({ default: undefined })).toBeNull()
+    expect(unwrapPluginDefault(null)).toBeNull()
+    expect(unwrapPluginDefault(undefined)).toBeNull()
+  })
+
+  it('returns null (never throws) when the default getter throws', () => {
+    const broken = { default: () => { throw new Error('boom') } }
+    expect(unwrapPluginDefault(broken)).toBeNull()
+  })
+
+  it('returns null when the default getter resolves to a non-object value', () => {
+    // Defensive: a misbehaving wrapper could resolve to undefined/null
+    expect(unwrapPluginDefault({ default: () => undefined })).toBeNull()
+    expect(unwrapPluginDefault({ default: () => null })).toBeNull()
+  })
+})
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Small, well-contained fix for desktop plugin discovery: `discoverBundledPlugins` now unwraps Rolldown's CJS-interop default-export getter so bundled plugins don't get silently skipped under the Rolldown bundler.

**What it does**
- `apps/desktop/src/contrib/plugins.ts` adds `unwrapPluginDefault(mod)` that accepts both the normal ESM shape (`{ default: plugin }`) and the Rolldown CJS-interop wrapper (`{ default: () => plugin }`); a throwing getter returns `null` (never throws at discovery).
- `import.meta.glob` type widened to `HermesPlugin | (() => HermesPlugin)`; `discoverBundledPlugins` calls the unwrap helper.
- New `plugins.test.ts` covers: ESM object passthrough, getter invocation, missing/undefined default → null, throwing getter → null, and getter resolving to non-object → null.

**Non-blocking notes**
- `unwrapPluginDefault` invokes the getter eagerly at discovery; the id/`register` validation still lives in the caller, so a valid object that isn't a plugin is skipped with the existing warning — good separation.
- The unwrap only inspects `default`; if a future bundler emits a named-export shape (`{ plugin }`) it returns `null` and the plugin is skipped with a warning. Fine as a defensive boundary, but the fallback behavior is worth keeping in mind.
- No other callers of `mod.default` for bundled plugins appear affected; the one-shot init guard already prevents re-discovery.

No functional issues found. Minimal, defensive, well-tested.

PR #97552, opened 2026-08-29
fix(desktop): route approvals by event owner scope

### LINKED ISSUE #97511: fix(desktop): route approval responses through exact runtime event owner

wesleysimplicio:
Closes #97511\n\nRemember the exact connection/profile that delivered a runtime event and use it as the fail-closed owner fallback for approval responses. Durable tile, hint, and stored-row ownership remains higher precedence.

### DIFF
--- a/apps/desktop/src/store/session-states-runtime-map.test.ts
+++ b/apps/desktop/src/store/session-states-runtime-map.test.ts
@@ -9,6 +9,7 @@ import {
   clearAllSessionStates,
   knownOwnerForSession,
   publishSessionState,
+  recordSessionEventScope,
   requestForOwnedSession,
   storedSessionIdForRuntimeId
 } from '@/store/session-states'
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #97519. Both use the inbound runtime event's exact connection/profile as the final approval-owner fallback while preserving durable owner precedence.

wesleysimplicio:
This behavior is already covered by #97519, which contains the same event-owner fallback plus the broader regression matrix and maintainer-facing explanation. To avoid keeping two PRs for the same fix, I am closing this narrower duplicate in favor of #97519.

PR #97556, opened 2026-08-29
feat(doctor): report optional skill scanner readiness

### LINKED ISSUE #97485: feat(skills): surface missing SkillEvaluator/SkillSpector and offer setup guidance
## Summary

Hermes supports NVIDIA SkillEvaluator Tier 1 and SkillSpector as an optional advisory layer, but the default configuration can say `skills.tier1_advisory: true` while neither scanner binary is installed. In that state the advisory scan is silently skipped.

Please surface scanner readiness in `hermes doctor` or `hermes tools`, and provide a clear one-command or copy-paste remediation path.

## Current behavior

- `skills.tier1_advisory` defaults to `true`.
- If `skillevaluator` is absent from `PATH`, the Tier 1 scan silently becomes a no-op.
- If SkillEvaluator is present but `skillspector` is absent, the security check is incomplete and is only visible when that install-time scan runs.
- A normal health check does not make it obvious that the configured advisory layer is unavailable.

This makes an enabled setting easy to mistake for an active security control.

## Proposed behavior

1. Have `hermes doctor` and/or `hermes tools` report one of:
   - Tier 1 advisory disabled
   - Tier 1 advisory ready: SkillEvaluator and SkillSpector available
   - Tier 1 advisory incomplete: list the missing binary or binaries
2. Show the pinned remediation commands already documented by Hermes, or add an explicit command such as `hermes skills security setup` that installs the optional scanners only after user confirmation.
3. Verify the binaries after setup and print the resulting readiness state.
4. Do not silently auto-install third-party scanners.

## Why this is useful

- Makes the effective security posture match what the configuration appears to promise.
- Avoids users discovering the missing scanner only after reading source or install-path documentation.
- Keeps the current advisory, fail-open design while making the degraded state observable.

## Environment observed

- Hermes Agent v0.20.6
- macOS
- `skills.tier1_advisory: true`
- `skillevaluator`: not installed
- `skillspector`: not installed

## Related implementation

- #88573
- SkillSpector integration commit: 2c2697b52e7ce608eb9236877231468be66bf915

wesleysimplicio:
Closes #97485\n\nMake the configured Tier 1 skill advisory posture visible in `hermes doctor`: disabled, ready, or incomplete with the missing optional scanner names and pinned install guidance. No third-party scanner is auto-installed.

### DIFF
--- a/hermes_cli/doctor.py
+++ b/hermes_cli/doctor.py
@@ -3169,6 +3169,31 @@ def run_doctor(args):
             issues.append("Run 'hermes setup' to configure missing API keys for full tool access")
     except Exception as e:
         check_warn("Could not check tool availability", f"({e})")
+
+    _section("Skill Security Advisory")
+    try:
+        from tools.skillevaluator_scan import scanner_readiness
+
+        readiness = scanner_readiness()
+        if readiness["state"] == "disabled":
+            check_ok("Tier 1 advisory disabled")
+        elif readiness["state"] == "ready":
+            check_ok("Tier 1 advisory ready", "(SkillEvaluator + SkillSpector)")
+        else:
+            missing = ", ".join(readiness.get("missing") or [])
+            check_warn("Tier 1 advisory incomplete", f"(missing {missing})")
+            check_info(
+                "Install the optional scanners with: "
+                "uv tool install --python 3.13 "
+                "'skillevaluator @ git+https://github.com/NVIDIA/SkillEvaluator.git@v0.1.0' "
+                "and `uv tool install 'git+https://github.com/NVIDIA/SkillSpector.git@v2.9.5'`."
+            )
+            issues.append(
+                "Tier 1 skill advisory is incomplete; install the missing "
+                "SkillEvaluator/SkillSpector scanner(s) or disable skills.tier1_advisory"
+            )
+    except Exception as e:
+        check_warn("Could not check skill advisory readiness", f"({e})")
     
     _section("Skills Hub")
     hub_dir = HERMES_HOME / "skills" / ".hub"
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Nice additive feature: `hermes doctor` now reports readiness of the optional Tier 1 skill-advisory scanner layer (SkillEvaluator + SkillSpector) via a new cheap, read-only `scanner_readiness()` PATH probe, with actionable install guidance when incomplete.

**What it does**
- `tools/skillevaluator_scan.py` adds `scanner_readiness()`: returns `{"state": "disabled"}` when `tier1_advisory_enabled()` is false (without probing), else `"ready"` or `"incomplete"` plus a `missing` list based on `shutil.which` for the two scanner binaries. Explicitly no install and no scan.
- `hermes_cli/doctor.py` adds a "Skill Security Advisory" section: `check_ok` for disabled/ready, `check_warn` + `check_info` (with the exact `uv tool install` commands) + `issues.append` when incomplete.
- Tests cover disabled (no `which` call), one-missing, and both-present cases.

**Non-blocking notes**
- The install commands (with pinned versions `@v0.1.0`, `@v2.9.5`) are hardcoded in the doctor message. If the canonical install path ever lives in a helper (e.g. a setup command), the message can drift from it. Consider sourcing the commands from one place if they exist elsewhere.
- `missing` uses the binary names as reported — on Windows, `uv tool install` shims land in `%APPDATA%\...\bin`, which `shutil.which` should find if on PATH; worth a quick manual check that the Windows path case reports correctly.
- Appending an issue when incomplete is a deliberate visibility choice; note it will count toward any doctor exit-code/summary that aggregates issues, which is appropriate for a security advisory.

No functional issues found. Well-tested, non-invasive addition.

PR #97558, opened 2026-08-29
fix(cron): treat a live multiplexer as gateway-alive for satellite profiles

### LINKED ISSUE #97120: Cron warning recommends `hermes gateway install` under multiplex; the gateway then refuses to start and the service crash-loops
## Summary

Under `gateway.multiplex_profiles: true`, creating a cron job in a **secondary** profile prints:

```
⚠  Gateway is not running — jobs won't fire automatically.
   Start it with: hermes gateway install
```

Both halves are wrong, and following the advice creates a launchd/systemd crash loop:

1. **The warning is a false negative.** The multiplexer *does* tick that profile's cron store.
2. **The remedy it recommends is forbidden by the gateway's own start guard**, which correctly refuses to start a per-profile gateway while the multiplexer is running — so the installed service exits `78` (`EX_CONFIG`) on every launch and the supervisor restarts it forever.

Two parts of the same codebase give contradictory instructions. Neither the cron liveness probe nor its remedy is multiplex-aware.

## Reproduction

1. Enable the multiplexer on the default profile: `gateway.multiplex_profiles: true`
2. Start the default gateway. It logs, correctly:
   `Cron scheduler will tick N profile(s) under multiplex: ['default', 'other', ...]`
3. In a secondary profile: `hermes -p <profile> cron create ...`
4. Observe the warning above.
5. Follow it: `hermes -p <profile> gateway install`
6. The service is installed and immediately starts failing:

```
✗ The default gateway is running as a profile multiplexer and already serves profile '<profile>'.
  When gateway.multiplex_profiles is on, the default gateway is the
  single inbound process for every profile. Starting a separate
  gateway for this profile would double-bind its platforms (two
  pollers on one bot token, port conflicts).
```

`launchctl list` then shows the job with no PID and last exit status `78`. With `KeepAlive: true` it respawns continuously — observed **~1 attempt every 17 s, 59 rejections** before I removed the job. On macOS this also causes visible UI churn in the Hermes desktop app, because the constant process start/stop moves files that the change-watcher broadcasts on (`sessions.changed` / `platforms.changed`), so the app re-renders in a loop.

## Why the probe is wrong

`hermes_cli/cron.py:67` — `_builtin_gateway_liveness()` decides liveness from two profile-scoped signals:

- `is_gateway_runtime_lock_active()` — the runtime lock lives in the **profile's own home**
- `find_gateway_pids()` at `hermes_cli/cron.py:98`

and `find_gateway_pids()` (`hermes_cli/gateway.py:952`) documents its default explicitly:

> `all_profiles`: […] When `False` (default), only PIDs belonging to the current Hermes profile are returned.

Under multiplex the gateway process lives in the **default** home, so a secondary profile has neither a lock nor a PID of its own. Liveness collapses to `False`, and `_warn_if_gateway_not_running()` (`hermes_cli/cron.py:103`, message at `:123-126`) fires.

Meanwhile `_guard_named_profile_under_multiplexer()` (`hermes_cli/gateway.py:6034`, message at `:6130`) already knows about multiplex and refuses precisely what the cron warning just recommended.

## Evidence the warning is a false negative

While the per-profile gateway was crash-looping and had never once started, the secondary profile's cron job ran normally, driven by the multiplexer:

```
last_run_at   2026-08-28T10:35:46   last_status: ok
next_run_at   2026-08-28T10:45:00
cron/ticker_heartbeat in that profile's home: 8 s old
```

## Impact

Beyond the misleading message, this is actively harmful in agent setups. An agent with terminal access that creates a cron job reads the printed remedy and executes it — it has no way to know the advice contradicts the gateway's own guard. That is exactly how I hit this: an agent created a scheduled job, followed the instruction, and left a crash-looping service behind. A human might merely be confused; an agent acts on it.

## Suggested fix

Make the liveness probe multiplex-aware. When the *default* profile's config has `gateway.multiplex_profiles: true` and its gateway is alive, any profile served by that multiplexer should report the ticker as live. Roughly: resolve the default home, check the multiplexer flag, and if set, evaluate the runtime lock / PID there instead of (or in addition to) the current profile's home — mirroring the resolution `_guard_named_profile_under_multiplexer()` already performs.

Failing that, at minimum the remedy text should be conditional: under multiplex it should point at `hermes gateway restart` from the default profile, never at `hermes gateway install`.

## Environment

- Hermes Agent 0.20.6, `main @ 3f315e46fe`
- macOS (Darwin 25.6.0), arm64, launchd-managed gateway
- CPython 3.11.15
- `gateway.multiplex_profiles: true` on the default profile, 6 profiles served

686f6c61:
## What does this PR do?

Under `gateway.multiplex_profiles: true`, a satellite profile has no `gateway.pid` of its own. Cron liveness only looked at the current profile's lock and PIDs, so `hermes -p other cron create` printed "Gateway is not running — start it with: hermes gateway install" even while the default multiplexer was already ticking that profile's job store.

Following the advice installs a named-profile service that the start guard correctly refuses (exit 78), which then crash-loops under KeepAlive.

Cron liveness now treats a live default multiplexer that serves this named profile as alive. The start-guard probe is shared (`named_profile_served_by_running_multiplexer`) so the two sites cannot disagree.

## Related Issue

Fixes #97120

## Type of Change

- [ ] Security fix
- [ ] Documentation update

## Changes Made

- `hermes_cli/gateway.py`: `named_profile_served_by_running_multiplexer()`; start guard uses it
- `hermes_cli/cron.py`: `_builtin_gateway_liveness` returns True when that probe is true
- `tests/cron/test_87033_cronjob_gateway_liveness.py`, `tests/gateway/test_multiplex_lifecycle.py`

## How to Test

```text
python -m pytest tests/cron/test_87033_cronjob_gateway_liveness.py tests/gateway/test_multiplex_lifecycle.py -q
# 26 passed
```

Not runtime-tested against a live launchd multiplex install.

### Code

- [x] I've run the targeted pytest files above

## Screenshots / Logs

N/A

### DIFF
--- a/hermes_cli/cron.py
+++ b/hermes_cli/cron.py
@@ -93,9 +93,16 @@ def _builtin_gateway_liveness() -> Optional[bool]:
             # scan below still decide instead of collapsing the whole
             # tri-state to None.
             pass
-        from hermes_cli.gateway import find_gateway_pids
+        from hermes_cli.gateway import (
+            find_gateway_pids,
+            named_profile_served_by_running_multiplexer,
+        )
 
-        return bool(find_gateway_pids())
+        if find_gateway_pids():
+            return True
+        # Satellite profile: no local gateway.pid, but the default multiplexer
+        # ticks this profile's cron store (#97120).
+        return named_profile_served_by_running_multiplexer()
     except Exception:
         return None
 
--- a/hermes_cli/gateway.py
+++ b/hermes_cli/gateway.py
@@ -6031,6 +6031,82 @@ def _running_under_gateway_supervisor() -> bool:
     return is_gateway_supervisor_process()
 
 
+def named_profile_served_by_running_multiplexer() -> bool:
+    """True when a live default multiplexer already ticks this named profile.
+
+    Shared by the named-profile start guard and cron liveness: a satellite
+    profile has no gateway.pid of its own, but the default multiplexer's
+    ticker still fires its jobs (#97120).
+    """
+    try:
+        suffix = _profile_suffix()
+    except Exception:
+        return False
+    if not suffix:
+        return False
+
+    try:
+        from hermes_constants import get_default_hermes_root
+        default_root = get_default_hermes_root()
+    except Exception:
+        return False
+
+    try:
+        from gateway.status import _read_pid_record
+
+        default_pid_path = default_root / "gateway.pid"
+        rec = _read_pid_record(default_pid_path)
+        if not rec:
+            return False
+        from gateway.status import _pid_exists, _pid_from_record
+        pid = _pid_from_record(rec)
+        if not pid or not _pid_exists(pid):
+            return False
+
+        from gateway.config import _env_multiplex_profiles_override
+
+        cfg_path = default_root / "config.yaml"
+        cfg = {}
+        if cfg_path.exists():
+            from hermes_cli.config import read_user_config_raw
+
+            cfg = read_user_config_raw(cfg_path)
+
+        env_multiplex = _env_multiplex_profiles_override()
+        if env_multiplex is False:
+            return False
+        if env_multiplex is True:
+            multiplex = True
+        else:
+            if not cfg_path.exists():
+                return False
+            multiplex = bool(
+                cfg.get("multiplex_profiles")
+                or (cfg.get("gateway", {}) or {}).get("multiplex_profiles")
+            )
+        if not multiplex:
+            return False
+
+        gateway_cfg = cfg.get("gateway", {}) or {}
+        if "multiplex_profile_allowlist" in cfg:
+            raw_allowlist = cfg.get("multiplex_profile_allowlist")
+        else:
+            raw_allowlist = gateway_cfg.get("multiplex_profile_allowlist")
+        from gateway.config import _normalize_multiplex_profile_allowlist
+        from hermes_cli.profiles import normalize_profile_name
+
+        profile_allowlist = _normalize_multiplex_profile_allowlist(raw_allowlist)
+        if (
+            profile_allowlist is not None
+            and normalize_profile_name(suffix) not in profile_allowlist
+        ):
+            return False
+        return True
+    except Exception:
+        logger.debug("Multiplexer-serving probe failed", exc_info=True)
+        return False
+
+
 def _guard_named_profile_under_multiplexer(force: bool = False) -> None:
     """Refuse a named-profile gateway when a multiplexer is already serving it.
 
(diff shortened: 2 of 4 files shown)


Enough1122:

**Overall:** Good extraction that fixes a cron liveness false-negative for satellite profiles: the named-profile-under-multiplexer probe logic becomes a reusable `named_profile_served_by_running_multiplexer()` boolean, shared by the start guard (unchanged behavior) and cron liveness (a live default multiplexer now counts as gateway-alive for the satellite profile, since it ticks that profile's cron store per #97120).

**What it does**
- `hermes_cli/gateway.py` extracts `named_profile_served_by_running_multiplexer()` returning `bool` (was the guard's inline probe, previously fail-open via the enclosing except → now fail-closed `False` on any probe error, debug-logged). `_guard_named_profile_under_multiplexer` keeps its exact prior semantics by calling the new function.
- `hermes_cli/cron.py:_builtin_gateway_liveness` — after `find_gateway_pids()` returns false, falls back to the multiplexer probe; the whole function still returns `None` (unknown) on exception, preserving the tri-state.
- Tests: existing guard tests updated to patch the new function; new tests assert a running multiplexer counts as alive for cron, no multiplexer + no pids is still false, and the served-probe matches the start guard.

**Non-blocking notes**
- The extracted function's config read mirrors the old precedence (env `GATEWAY_MULTIPLEX_PROFILES` override, then `config.yaml`), but note `env_multiplex is False` now returns `False` — i.e. "multiplexer not serving". For cron liveness that means an explicitly multiplexer-disabled satellite must be covered by its own `find_gateway_pids()` path, which is correct.
- Fail-closed behavior change: the old inline probe propagated `None` (unknown) upward when it threw; now the sub-probe returns `False` and cron returns `None` only via its own enclosing try. The net liveness tri-state is preserved (verified by `test_lock_probe_failure_still_falls_back`).
- The extraction duplicates the multiplex-config parsing in one place now (single function) — good; just be aware the `_guard_...` docstring/comment volume is high relative to the logic.

No functional issues found. Refactor is behavior-preserving for the guard and fixes the cron liveness gap with good coverage.

teknium1:
Merged via #99441 (commit e703717513) — cherry-picked onto current main with your authorship preserved, zero conflicts. Your read of the polarity was exactly right: #99375's fix killed the false-positive, and your shared `named_profile_served_by_running_multiplexer()` extraction killed the false-negative while guaranteeing cron and the start guard can never disagree. The exit-78 crash-loop path (#97120) is gone. Thank you @686f6c61!

PR #97559, opened 2026-08-29
fix(tools): propagate database errors in session_search instead of masking as not-found

shali10:
### Summary

In `session_search` (both `read` and `scroll` shapes), `db.get_session()` was wrapped in a broad try/except block that logged only at `DEBUG` level and swallowed any exception into an empty `meta = {}`, immediately leading to `tool_error(f"session_id not found: {session_id}")`.

This caused transient storage issues (e.g. SQLite locks, disk I/O errors, pool contention) to be misreported as "session does not exist". In production, this misleads agents into believing a valid session was deleted and triggers unnecessary fallbacks.

### Changes
- In `_read_session` and `_scroll`, catch unexpected `Exception` from `db.get_session()`, log at `ERROR` level, and return `tool_error(f"failed to load session: {e}", success=False)` to clearly distinguish DB failures from actual missing sessions.
- Add regression test cases in `tests/tools/test_session_search.py`.

### DIFF
--- a/tests/tools/test_session_search.py
+++ b/tests/tools/test_session_search.py
@@ -10,6 +10,7 @@ All run zero LLM calls.
 """
 import inspect
 import json
+import sqlite3
 import time
 
 import pytest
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Correct fix: `session_search`'s read and scroll paths now propagate real database errors (`tool_error("failed to load session: ...")`, `success: false`) instead of masking them as "session_id not found", so transient DB failures are no longer indistinguishable from genuinely missing sessions.

**What it does**
- `tools/session_search_tool.py` — in `_read_session` and `_scroll`, `db.get_session(...)` exceptions now log at `error` and return `tool_error(f"failed to load session: {e}", success=False)`; the `or {}` fallback is removed so a `None`/empty result still hits the "session_id not found" branch.
- Tests add `TestSessionSearchErrorPropagation` for both read and scroll shapes (sqlite `OperationalError` surfaced with the message in the tool error).

**Non-blocking notes**
- The raw exception text is embedded in the tool error string — appropriate for tool output (the model needs the cause), and it already includes the message from the sqlite error. Fine as-is.
- Logging was upgraded from `debug` to `error` — a noisy-but-rare path (only on DB failures), reasonable. If this fires on routine lock contention (e.g. concurrent writers), the log volume may warrant a warning-level note; current choice is defensible.
- The `meta = db.get_session(session_id)` without `or {}` correctly distinguishes `{}`/`None` (not found) from a raised exception (real failure) — the two test shapes pin both.

No functional issues found. Minimal, well-targeted fix.

PR #97560, opened 2026-08-29
fix(compression): cancel lean digests before session reset

wesleysimplicio:
## Summary
- stop a cancelled lean compaction before it starts another chunk digest request
- retain the session compression lease through total-ceiling worker unwind to prevent overlapping retries
- report total-ceiling timeouts accurately
- classify cooldown-blocked context-overflow recovery as a soft deferred result instead of permanent exhaustion

## Validation
- `python -m py_compile agent/context_compressor.py agent/conversation_compression.py agent/conversation_loop.py run_agent.py`
- `python -m pytest -q tests/agent/test_compression_cancellation_97488.py tests/agent/test_compression_stall_fallback_78981.py tests/agent/test_compress_context_progress_timeout.py -k 'not owned_timeout_skips_hung_compress and not fallback_prompt_resolved_lazily_on_timeout and not caller_fence_bypasses_owned_wrapper'`
- Result: 26 passed; 3 existing environment-blocked tests require `python-dotenv`.

### DIFF
--- a/agent/context_compressor.py
+++ b/agent/context_compressor.py
@@ -1197,6 +1197,24 @@ def _digest_worthy(role: str, content: str) -> bool:
     return True
 
 
+def _compression_cancelled(compressor: Any) -> bool:
+    """Return whether the host has cancelled this compression attempt.
+
+    The progress-aware host cannot kill a provider request that is already in
+    flight, but it can prevent the next request in a multi-call compaction
+    plan. Keep this check callback-based so detached workers do not need to
+    share mutable host state beyond the commit fence.
+    """
+    check = getattr(compressor, "_compression_cancelled_check", None)
+    if not callable(check):
+        return False
+    try:
+        return bool(check())
+    except Exception:
+        logger.debug("compression cancellation check failed", exc_info=True)
+        return False
+
+
 def _serialize_turns_for_digest(
     turns: List[Dict[str, Any]],
     pristine: "dict[str, str] | None" = None,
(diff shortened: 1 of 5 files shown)


alt-glitch:

Related to #97512 and #93241. This version strengthens cancellation after a total-ceiling timeout and classifies cooldown/lock deferral to avoid a false session reset.

kchuang1015 (commented):
Independent verification on exact head `c988aaf73c98a13469deb3bdcc794980dd5c0ae9`:

- Canonical focused command with file retries disabled: **29 passed, 0 failed**. This includes the three tests that the PR description says were environment-blocked locally.
- Adjacent compression/gateway matrix: **255 passed, 1 deselected**. The deselected S3 elapsed-time test is host-startup sensitive here and also failed 5/5 on an unmodified current `main` worktree; it is not introduced by this diff. The exact-head GitHub Linux lane is green at **39,793 passed, 0 failed, 383 skipped**.
- Synthetic orchestration RED/GREEN: a structured provider context-overflow followed by an unchanged compression result under an active cooldown returns `compression_exhausted` on current `main`, but returns `compression_deferred=True`, `failed=False`, and no exhaustion on this head. The temporary fixture was removed after execution.
- `git diff --check`, AST parsing of all five changed Python files, and a high-confidence diff secret/path scan passed with zero findings.

One coverage request before merge: `test_active_cooldown_is_a_soft_overflow_defer` currently calls `compression_deferred_reason()` and `_compression_deferred_result()` directly, so it does not prove the three modified orchestration sites in `agent/conversation_loop.py` actually use the result. Please promote the synthetic RED/GREEN shape above into a committed end-to-end regression, modeled on `tests/run_agent/test_compression_lock_defer.py`: make the client raise a structured context overflow, make `_compress_context()` return the same list while `get_active_compression_failure_cooldown()` reports a live cooldown, then assert `compression_deferred`, no `compression_exhausted`, and `failed=False`. The negative control fails on current `main` for the intended reason.

teknium1:
PR #98628 (merged, 4f22543509) removed the per-chunk lean digest loop entirely — lean compaction is now a single auxiliary request, so there are no in-flight digest calls left to cancel before a session reset. Closing as overtaken; the surviving general concern (canceling a live compression attempt on reset) is tracked by #97488. Thanks for the work — the cancellation-ordering analysis was sound against the code as it stood.

Issue #97178, opened 2026-08-28
[Bug]: Desktop Settings → Model memory toggles write to the wrong profile — flips cascade across all profiles


jackthelobster:
### Bug Description

In the **desktop app → Settings → Model** panel, the **Persistent memory** and **User profile** (memory) toggles do not respect the selected profile. Flipping either toggle appears to apply to **every profile at once**, and flipping it back cascades the same way — there is no way to scope the setting to a single profile from the UI.

Observed with 5 profiles (`default`, `coach`, `hdops`, `ponytail`, `txops`), each an independent `HERMES_HOME` under `~/.hermes/profiles/`:

1. User enabled persistent memory + user profile for all profiles **except** ponytail via the desktop settings toggles.
2. Result: **every** profile had the settings flipped on, including ponytail.
3. User then toggled memory **off** for ponytail.
4. Result: **every** profile flipped off.

### Expected behavior

The toggles should write `memory.memory_enabled` / `memory.user_profile_enabled` to the **selected profile's** `config.yaml` (that profile's `HERMES_HOME`), as the dashboard's profile switcher implies.

### Actual behavior

The toggle appears to write to a single (wrong) scope — most likely the desktop app process's own `HERMES_HOME` — so all profiles read the change as their own, and per-profile state cannot be set from the UI. This matches the wrong-scope write pattern already reported for the web dashboard (#92520, #79655) and multiplex slash commands (#75684), but on the desktop app Settings → Model surface.

### Environment

- Hermes Agent v0.20.6 (2026.8.27), upstream 9978706e, local 0fa78369 (+2 carried commits)
- Install method: git, Termux/Android (aarch64), Python 3.13.13
- 5 profiles under `~/.hermes/profiles/`, desktop app, `gateway.multiplex_profiles: true` on one profile

### Workaround (verified working)

Set each profile's config via the CLI with `HERMES_HOME` pointed at the profile directory:

```bash
# enable for a profile
HERMES_HOME=~/.hermes/profiles/hdops hermes config set memory.memory_enabled true
HERMES_HOME=~/.hermes/profiles/hdops hermes config set memory.user_profile_enabled true

# explicitly disable for the one profile that must stay off
HERMES_HOME=~/.hermes/profiles/ponytail hermes config set memory.memory_enabled false
HERMES_HOME=~/.hermes/profiles/ponytail hermes config set memory.user_profile_enabled false
```

[...]

PR #97563, opened 2026-08-29
fix(desktop): preserve model profile in request body

wesleysimplicio:
Related to #97515

## Summary
- include an explicit settings profile in the model assignment JSON payload
- keep the existing IPC/profile routing metadata unchanged
- add a regression assertion for shared-remote model assignments

When a shared remote normalizes the Electron request envelope to its primary backend, the backend can still honor `body.profile` and scope the config write to the selected profile.

This is complementary to #97530, which addresses the broader shared-remote routing path used by reads and sibling settings pages.

## Validation
- `git diff --check`
- Desktop Vitest could not start in this checkout because `vitest/config` is unavailable in the installed dependencies.

### DIFF
--- a/apps/desktop/src/api/models.ts
+++ b/apps/desktop/src/api/models.ts
@@ -120,10 +120,16 @@ export function setModelAssignment(
   body: ModelAssignmentRequest,
   profile?: null | string
 ): Promise<ModelAssignmentResponse> {
+  // Keep the target in the JSON payload as well as the IPC request envelope.
+  // A shared remote can route the request to its primary backend while the
+  // envelope's profile is being normalized; the backend's model handler still
+  // honors body.profile when choosing the config home (#97515).
+  const requestBody = profile == null ? body : { ...body, profile }
+
   return hermesApi<ModelAssignmentResponse>({
     ...profileScoped(profile),
     path: '/api/model/set',
     method: 'POST',
-    body
+    body: requestBody
   })
 }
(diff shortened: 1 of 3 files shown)


alt-glitch:

Related to #97530 and #97515. #97530 changes the broader Desktop settings request-scope contract; this PR adds a model-set payload-level safeguard for shared remotes.

zakhounet:
Thanks for the quick PR. One caveat from the reporter's side: our packet capture shows the profile is dropped **generically** in the Electron main-process routing (the reused primary descriptor lacks `sharedRemote`, so `pathForRegistryBackendRequest` never adds `?profile=`). That drop affects **every config-backed read and write** on a shared-remote primary — the reads are why all profiles display the launch profile's values, and the same applies to Tools & Keys, Messaging, and the memory toggles (#97178).

`body.profile` fixes the write on this one endpoint, which is a good belt-and-braces for `/api/model/set`, but merging this alone would auto-close #97515 with the read path and the sibling pages still broken. #97530 addresses the routing itself (`sharedRemote: true` on the reused descriptor) — we've verified that exact change locally on two machines.

Maybe treat this PR as complementary and drop the `Closes` so the issue tracks the full fix?

wesleysimplicio:
Thanks for the detailed report. Agreed: this payload safeguard is complementary, not the complete shared-remote fix. I removed the premature `Closes #97515` linkage and documented #97530 as the broader routing fix, so the issue remains open for the full read/write path.

PR #95432, opened 2026-08-26
fix(cron): fence contention no longer reads as fire-claim ownership loss

BrunoBza:
## What does this PR do?

Fixes #95307 — finite cron jobs (once / repeat-limited) delivering to `bot-chat` complete their agent turn and then get recorded as failed with `Fire claim ownership lost; stale result was discarded.`, losing the result entirely. The root cause is a semantic conflation in the fire-claim ownership probe; this PR separates "definitively lost" from "could not verify" so the runner stops discarding its own successful result.

## Bug cause

`heartbeat_fire_claim()` returned `False` for **two different facts**:

1. the claim is gone or now belongs to another owner (a definitive takeover), and
2. the per-job fire fence could not be acquired within `\_JOBS_LOCK_TIMEOUT_SECONDS` — pure lock contention, with **no information** about who owns the claim.

Those are not interchangeable. The fence that blocks the refresh is the very lock that *excludes replacement owners* while the legitimate runner performs a fenced side effect. A `bot-chat` delivery runs an entire child-agent turn (`hermes chat -Q ...`) while holding that fence, so heartbeat/probe calls issued during delivery contend with the runner's own fence. Reading that as ownership loss makes the runner discard its own completed result. Deliveries to `local` / ordinary platforms finish in well under a second and never straddle the probe window, which matches the reporter's isolation matrix exactly.

On the current main, `mark_job_run(expected_fire_owner=...)` already CAS-guards terminal writes and the post-run probes run before finalization — so the reporter's proposed fixes #1/#2 are largely landed upstream. What remains broken is exactly the conflated `False`: any caller that cannot acquire the fence is told "you lost the claim" when the truth is "I don't know".

## Fix

`heartbeat_fire_claim()` becomes tri-state:

| Result | Meaning |
|---|---|
| `True` | claim exists, belongs to `expected_owner`, refreshed |
| `False` | definitive loss: job/claim gone, or owner replaced |
| `None` | fire fence busy — ownership unknown, no evidence either way |

Scheduler consumers updated accordingly:

- **Pre-run validation**: `None` closes the execution row with the existing "Fire claim ownership could not be validated before execution started." record (same treatment as a transport exception) instead of reporting a loss.
- **Heartbeat loop**: a `None` beat is skipped without consuming the exception grace budget; refreshes resume once the fence frees up. Only a definitive `False` sets the ownership-lost event.
[...]

Issue #95307, opened 2026-08-26
[Bug]: cron — result of finite jobs (once / repeat-limited) discarded as "Fire claim ownership lost" when bot-chat delivery is configured


alex-frolov:
## Bug Description

Cron jobs whose **dispatch is finite** (schedule kind `once`, or a cron-expression job created with `--repeat N`) or whose **delivery target list contains `bot-chat`** complete their agent turn successfully but the result is then discarded and **no delivery happens** (neither Telegram nor Bot Chat). `hermes cron runs` records:

```
failed     Fire claim ownership lost; stale result was discarded.
```

The model actually runs every time (tokens appear in `~/.hermes/cron/usage_audit.jsonl`), so the failure is purely in the post-run ownership/finalization path, not in execution.

Environment detail: the desktop app's built-in ticker (`HERMES_DESKTOP=1`) is the firing process in all reproductions. An identical one-shot fired by a standalone `hermes cron tick` process in an isolated `HERMES_HOME` completes `ok` — the defect needs the desktop ticker's finalization path.

Isolation matrix (all on v0.20.5, macOS, single profile, gateway launchd service running alongside the desktop app — ticks serialized by `.tick.lock`):

| # | job shape | `--deliver` | result |
|---|---|---|---|
| A | `once @ ISO` | `bot-chat,telegram:<id>` | ✗ discarded |
| B | `once @ ISO` | `bot-chat,telegram:<id>` | ✗ discarded |
| C | cron-expr date + `--repeat 1` | `bot-chat,telegram:<id>` | ✗ discarded |
| D | cron-expr date, no repeat | `bot-chat,telegram:<id>` | ✗ discarded |
| E1 | cron-expr date, no repeat | `local` | ✓ completed |
| E2 | cron-expr date, no repeat | `telegram:<id>` | ✓ completed |
| F | cron-expr date, no repeat | `local,telegram:<id>` | ✓ completed |
| — | recurring `*/2 * * * *` | `local` | ✓ completed |

Conclusion: two independent triggers, same failure mode —

1. any finite-dispatch job (`once`, or repeat-limited via `claim_dispatch`),
2. …combined with `bot-chat` present in the delivery target list.

## Steps to Reproduce

On a machine where the Hermes desktop app is running (desktop cron ticker active):

```bash
# fires in ~2 minutes; completes its agent turn, then discards the result
hermes cron create "$(date -v+2M '+%M %H %d %m') *" \
  "Reply with exactly: OK" \
[...]

Issue #88688, opened 2026-08-17
Cron/session recovery: fence reconciliation by exact ownership generation


andrexibiza:
## Summary

The recent session-persistence and cron-recovery train closes several real data-loss and silent-wedge paths, but the combined architecture still has one shared correctness gap: **recovery operations are not consistently fenced against a successor owner**.

The relevant fixes are individually sound for the incidents they address:

- #88244 prevents CLI cleanup from finalizing a session after `/handoff` has transferred it to the gateway, and surfaces `state.db` initialization failure instead of silently degrading persistence.
- #88335 stops `EMFILE`/`ENFILE` from being classified as harmless tick-lock contention and adds degraded-health/backoff behavior.
- #88337 clears one-shot `run_claim` on pre-dispatch failures.
- #88339 re-arms recurring jobs wedged in restart-surviving persisted error state.
- #88343 reconciles stale process-local in-flight claims against the durable execution ledger.

Taken together, however, these fixes reveal the next class boundary: **an old recovery actor can still clear, release, finalize, or re-arm state that now belongs to a newer generation of the same session/job.**

The invariant should be:

> Every attach, clear, finalize, release, or reconcile operation must identify and conditionally match the exact ownership generation it intends to mutate. A stale actor must be unable to mutate its successor.

This issue tracks that class as one coherent recovery/fencing problem rather than another series of incident-local TTL guards.

## 1. `/handoff`: completion-to-cleanup still has a narrow ownership race

#88244 adds the session ID to `_handed_off_session_ids` only after the CLI polls `get_handoff_state()` and observes `state == "completed"`. Cleanup suppression is therefore process-local knowledge acquired *after* the gateway has committed the handoff.

There is still a legal ordering:

1. Gateway completes the handoff and assumes/reopens the session.
2. CLI has not yet observed `completed`, so its session ID is not yet in `_handed_off_session_ids`.
3. CLI receives SIGTERM / abnormal shutdown / another cleanup trigger.
4. Old CLI finalizes the session now owned by the gateway.

That is the original ownership bug compressed into a smaller window.

Marking the session as handed off earlier is not sufficient: a failed handoff must leave the CLI as the legitimate owner. The transfer needs a durable generation/lease or equivalent CAS boundary:

```text
CLI owns session S generation A
Gateway prepares generation B
commit atomically transfers A -> B
CLI finalize(S, expected_generation=A) succeeds only while A still owns S
```

[...]

PR #97565, opened 2026-08-29
fix(cron): tri-state fire-claim heartbeat — lock timeout is 'unverifiable', not ownership loss

sycamoregroupltd:
## Problem

`heartbeat_fire_claim()` in `cron/jobs.py:3376` returns `False` in **two distinct situations**:

1. ownership is *positively observed* to have changed (claim absent / owned by another), and
2. the per-job fire fence could not be acquired within `_JOBS_LOCK_TIMEOUT_SECONDS` (30s) — a fail-closed "could not verify".

`cron/scheduler.py:6985` collapses both into `fire claim ownership lost; interrupting stale run` and **kills a healthy run**, which is then recorded as a cron FAILURE.

Under cgroup memory/CPU starvation (measured: PSI full avg10 26–37%, `memory.current == memory.high`, swap exhausted), a 30s lock wait is routine — so a single heartbeat tick can false-positive "ownership lost" and kill a run that was never stolen.

## Fix

Change the contract to **tri-state `Optional[bool]`**:

- `True` — claim still owned by `expected_owner` and refreshed
- `False` — ownership **positively observed** changed (claim absent/other owner)
- `None` — could **not verify** (fence lock unavailable) — NOT evidence of loss

Callers updated:

- **fire-claim heartbeat loop** (scheduler): `False` interrupts immediately; `None` feeds the existing `_FIRE_CLAIM_HEARTBEAT_GRACE_SECONDS` grace path (same as exceptions), so a run is interrupted only when ownership is positively observed changed **or** verification has failed continuously past the grace window.
- **initial validation**: `None` fails closed as `Fire claim ownership could not be validated...` (not "lost").
- **shutdown probes** (`_fire_claim_ownership_lost`, shutdown `mark_job_run` paths): only `True` counts as owned; `None` is not treated as loss.

## Verification

- 3 new regression tests in `tests/cron/test_script_claim_heartbeat.py`:
  - `test_unverifiable_fire_claim_not_killed_immediately` — the exact incident: None heartbeats must NOT interrupt a live run.
  - `test_unverifiable_fire_claim_cancels_after_bounded_grace` — sustained None past grace still cancels.
  - `test_initially_unverifiable_fire_claim_finishes_without_running` — initial None fails closed without claiming loss.
- Full `tests/cron/` suite: 1035 passed, 1 skipped. The single failure (`test_run_job_cron_execute_code_deny_does_not_pollute_later_gateway_execute_code`) is **pre-existing on `main`** (verified by stash) and unrelated (execute-code approval session isolation).

## Related

- Fleet incident: cron 'fire claim ownership lost' is a fail-closed lock timeout, not claim theft (2026-08-29) — `heartbeat_fire_claim()` returns False on lock timeout and the scheduler kills a healthy run.

### DIFF
--- a/cron/jobs.py
+++ b/cron/jobs.py
@@ -3373,10 +3373,25 @@ def _sweep_completed_oneshots(
     return removed
 
 
-def heartbeat_fire_claim(job_id: str, *, expected_owner: str) -> bool:
+def heartbeat_fire_claim(job_id: str, *, expected_owner: str) -> Optional[bool]:
+    """Confirm or refresh an active ``fire_claim`` for ``expected_owner``.
+
+    Returns tri-state so callers can distinguish a *positively observed* loss
+    from a transient failure to verify:
+
+    - ``True`` — the claim is still owned by ``expected_owner`` and was
+      refreshed.
+    - ``False`` — ownership is positively observed to have changed (the claim
+      is absent or belongs to another owner). Callers may interrupt the run.
+    - ``None`` — ownership could NOT be verified (the per-job fire fence could
+      not be acquired within ``_JOBS_LOCK_TIMEOUT_SECONDS``). This is NOT
+      evidence of loss: under memory/CPU starvation a healthy owner can fail
+      to reach the lock. Callers should treat ``None`` as "unknown" and feed
+      the existing grace path, not the immediate-kill path.
+    """
     with _fire_job_lock(job_id) as acquired:
         if not acquired:
-            return False
+            return None
         return _heartbeat_fire_claim_locked(
             job_id,
             expected_owner=expected_owner,
--- a/cron/scheduler.py
+++ b/cron/scheduler.py
@@ -6978,36 +6978,50 @@ def _run_with_fire_claim_heartbeat(job: dict, run) -> bool:
         _finish_unstarted("Fire claim ownership lost before execution started.")
         return True
 
+    if owns_fire_claim is None:
+        logger.warning(
+            "Job '%s': fire claim ownership could not be verified before execution; "
+            "aborting as unverifiable (not claiming loss)",
+            job_id,
+        )
+        _finish_unstarted(
+            "Fire claim ownership could not be validated before execution started."
+        )
+        return True
+
     def _heartbeat_loop() -> None:
         last_confirmed = time.monotonic()
         while not stop.wait(_RUN_CLAIM_HEARTBEAT_SECONDS):
             try:
-                if not heartbeat_fire_claim(job_id, expected_owner=owner):
-                    lost_ownership.set()
-                    logger.warning(
-                        "Job '%s': fire claim ownership lost; interrupting stale run",
-                        job_id,
-                    )
-                    return
-                last_confirmed = time.monotonic()
+                claim = heartbeat_fire_claim(job_id, expected_owner=owner)
             except Exception:
                 logger.debug(
                     "Job '%s': fire_claim heartbeat failed",
                     job_id,
                     exc_info=True,
                 )
-                if (
-                    time.monotonic() - last_confirmed
-                    >= _FIRE_CLAIM_HEARTBEAT_GRACE_SECONDS
-                ):
-                    lost_ownership.set()
-                    logger.warning(
-                        "Job '%s': fire_claim could not be renewed within %.1fs; "
-                        "interrupting uncertain run",
-                        job_id,
-                        _FIRE_CLAIM_HEARTBEAT_GRACE_SECONDS,
-                    )
-                    return
+                claim = None
+            if claim is False:
+                lost_ownership.set()
+                logger.warning(
+                    "Job '%s': fire claim ownership lost; interrupting stale run",
+                    job_id,
+                )
+                return
+            if claim is True:
+                last_confirmed = time.monotonic()
+            elif (
+                time.monotonic() - last_confirmed
+                >= _FIRE_CLAIM_HEARTBEAT_GRACE_SECONDS
+            ):
+                lost_ownership.set()
+                logger.warning(
+                    "Job '%s': fire_claim could not be renewed within %.1fs; "
+                    "interrupting uncertain run",
+                    job_id,
+                    _FIRE_CLAIM_HEARTBEAT_GRACE_SECONDS,
+                )
+                return
 
     heartbeat_thread = threading.Thread(
         target=heartbeat_context.run,
@@ -7134,8 +7148,7 @@ def _run_one_job_body(
         if fire_owner is None:
             return False
         try:
-            if heartbeat_fire_claim(job["id"], expected_owner=fire_owner):
-                return False
+            claim = heartbeat_fire_claim(job["id"], expected_owner=fire_owner)
         except Exception:
             logger.debug(
                 "Job '%s': fire_claim ownership validation failed",
@@ -7143,9 +7156,15 @@ def _run_one_job_body(
                 exc_info=True,
             )
             return False
-        if fire_claim_lost is not None:
-            fire_claim_lost.set()
-        return True
+        if claim is True:
+            return False
+        if claim is False:
+            if fire_claim_lost is not None:
+                fire_claim_lost.set()
+            return True
+        # claim is None — could not verify. Do NOT claim loss without a
+        # positive observation (same fail-closed lie as the heartbeat path).
+        return False
 
     execution_id = job.get("execution_id")
     if not execution_id:
@@ -7242,7 +7261,7 @@ def _run_one_job_body(
             # interruption through the owner-fenced terminal write.
             if fire_owner is not None and heartbeat_fire_claim(
                 job["id"], expected_owner=fire_owner,
-            ):
+            ) is True:
                 mark_job_run(
                     job["id"],
                     False,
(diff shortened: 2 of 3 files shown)


alt-glitch:

Related to #95432: both repair false fire-claim ownership loss under fence contention for #95307. This version routes an unverifiable heartbeat through the bounded grace path; #95432 skips that beat immediately.

andrexibiza (commented):
Reviewed exact head `474eb4ea1653e1defda9490d834b8b429ad5b23e` against live `main@ac6c8028e00d01ee2f299ba7fd03329c7f10382d`; this is currently a clean 1-commit/3-file branch directly on main. I read the actual patch plus the surrounding fire-fence, cancellation, delivery/finalization paths; the exact-head workflows; open #95307; direct predecessor #95432; and the complementary generation-fencing work tracked by #88688.

There is good work here. Splitting `heartbeat_fire_claim()` into owned / positively-lost / unverifiable is the right semantic direction. The initial `None` path now refuses to start without falsely claiming takeover, the shutdown probes stop using truthiness as ownership proof, and the hosted Python/e2e/macOS/Windows/lint lanes are green. The remaining blockers are at the exact boundary this PR is trying to repair.

**P1 — sustained `None` is still converted into the exact positive ownership-loss signal.**

In `cron/scheduler.py::_heartbeat_loop` (the hunk beginning around `@@ -6978`), `None` correctly enters the grace path, but after `_FIRE_CLAIM_HEARTBEAT_GRACE_SECONDS` that path calls `lost_ownership.set()`. The same event is then passed into `run_job()` as the fire-claim cancellation event. `run_job::_abort_if_fire_claim_lost()` consumes any set event by calling `agent.interrupt("Cron fire claim ownership was lost")` and raising `RuntimeError("... lost its durable fire claim ownership")`.

So the new state machine is only tri-state at the probe. At the execution sink it collapses back to two states:

`None for long enough -> lost_ownership event -> "ownership was lost"`.

That is not just wording. A healthy execution whose fence remains unavailable/contented through the grace window is still terminated through the same authority-loss channel as an observed successor owner. The incident class is therefore delayed rather than closed for long-running contention. The new `test_unverifiable_fire_claim_cancels_after_bounded_grace` currently codifies that collapse by asserting only that the loss event becomes set; it never asserts the exact terminal reason or exercises the real fire fence.

There is an important other side here before simply choosing “skip `None` forever.” `_fire_job_lock()` deliberately fails closed when the cross-process fire-fence backend is unavailable, and the same `acquired=False` surface also represents lock acquisition failure/contention. Those are not equivalent facts either. A busy fence held by the legitimate owner can itself be evidence that a replacement owner is excluded; a missing/broken cross-process locking backend is genuinely unverifiable. `Optional[bool]` still cannot carry that distinction.

**Required fix:** keep uncertainty distinct from observed loss end-to-end. If policy requires aborting after bounded unverifiability, use a separate typed cancellation/settlement reason (`unverifiable` / `indeterminate`, not `lost`) so no downstream path can claim takeover without a `False` observation. Better still, make the probe result explicit enough to distinguish at least `OWNED`, `LOST`, `FENCE_BUSY`, and `UNAVAILABLE/UNKNOWN`, because those states have different safe continuations. Add deterministic witnesses using the real fire-fence path: (1) legitimate fence contention that lasts across multiple heartbeat beats and past the configured grace, (2) unavailable locking backend, and (3) an actual owner replacement. Assert both effect suppression/continuation and the exact terminal reason. A real takeover must remain the only path that says ownership was lost.

**Merge/interlock — this is a second implementation of the same open defect class already carried by #95432.**

#95432 (BrunoBza) predates this PR, explicitly fixes reporter alex-frolov's #95307, changes the same `cron/jobs.py` + `cron/scheduler.py` surfaces, and introduces the same tri-state contract. Its semantic choice differs: a busy-fence `None` does not consume the exception grace budget, and its dedicated real-store regression suite exercises a long fenced delivery, post-delivery bookkeeping, startup contention, and a genuine takeover control. This PR is much fresher and is cleanly based on current main, but it should not silently become a parallel owner of the same repair.

Please converge these into one canonical landing object: preserve BrunoBza's implementation/test credit and alex-frolov's reporter/root-cause lineage, absorb the strongest real-fence witnesses, and make the contention-vs-backend-unavailable distinction explicit rather than choosing between two incomplete `None` policies. #88688 is complementary long-term work: exact ownership generation/CAS is the broader architecture that should eventually remove these ambiguous probes; it is not a duplicate of this incident repair.

**Acceptance blocker — exact head is not green.**

For this exact SHA:

- CI `33226106068`: **failure** — https://github.com/NousResearch/hermes-agent/actions/runs/33226106068
- `Check contributors / check-attribution` job `99030104601`: **failure** at “Check for unmapped contributor emails” — https://github.com/NousResearch/hermes-agent/actions/runs/33226106068/job/99030104601
- required-check aggregate `99030582376`: **failure** as a consequence — https://github.com/NousResearch/hermes-agent/actions/runs/33226106068/job/99030582376
- Docker `33226105599`: success — https://github.com/NousResearch/hermes-agent/actions/runs/33226105599
- Nix `33226105712`: success — https://github.com/NousResearch/hermes-agent/actions/runs/33226105712

The sole commit is authored as `Frank Spencer <frank@frankspencer.dev>`; the current tree has no mapping for that email, matching the attribution failure. Preserve Frank's authorship and add the repository contributor mapping rather than rewriting history, then require the corrected exact head to go fully green.

The core insight here is strong: “could not verify” is not “lost.” Carry that distinction all the way through cancellation and settlement, reconcile the earlier carrier instead of competing with it, and get the exact commit green. That closes the class instead of moving the false-positive boundary three minutes downstream. 🚀

jrodriguezruibal:
Confirming this diagnosis from a production gateway, and adding the piece I think is missing: the fail-closed "could not verify" branch is triggered routinely, not only under a same-job race.

**Environment:** single `hermes-gateway.service` (systemd user unit), Linux, ~14 cron jobs, checkout at `a0749d583a`, managed topology (agent jobs handed to `cron.scheduler --external-worker-file` transient scopes). Two independent events on the same agent-mode job (`1dd048646ab6`, every 30m, run time 3-11 min):

```
# event 1
13:17:15  Cron job '60db2e8eec47' handed to restart-safe worker pid=116668
13:17:44  ERROR cron.jobs: Timed out waiting for local fire fence <HERMES_HOME>/cron::1dd048646ab6; failing closed
13:17:44  WARNING __main__: Job '1dd048646ab6': fire claim ownership lost; interrupting stale run
13:18:46  Job '1dd048646ab6': delivered to Bot Chat of profile 'inbox-watch'
13:18:46  status=error, "Interrupted by shutdown before terminal completion."
```

```
# event 2
13:37:15  Cron job '945e877ca286' handed to restart-safe worker pid=122177
13:37:15  Cron job '1dd048646ab6' handed to restart-safe worker pid=122178
13:39:27  [cron_1dd048646ab6_20260910_133757] Turn ended: reason=text_response api_calls=13/90
13:40:45  ERROR cron.jobs: Timed out waiting for local fire fence <HERMES_HOME>/cron::1dd048646ab6; failing closed
13:40:45  WARNING __main__: Job '1dd048646ab6': fire claim ownership lost; interrupting stale run
13:41:34  Job '1dd048646ab6': delivered to Bot Chat of profile 'inbox-watch'
```

Both times the fence timeout and the ownership-lost line land in the same second, and the timeout is exactly `_JOBS_LOCK_TIMEOUT_SECONDS` (30s) after a *different* job's worker was spawned. The victim job was healthy and mid-run (heartbeating every 60s: 13:39:45 ok, 13:40:45 lost), it finished its turn, it delivered — and it was still recorded as `failed`.

**Triggers observed:** (a) two jobs firing in the same tick, each in its own worker (13:37:15 above); (b) a worker spawning for another job while this one is mid-run (13:17:15). In every case the victim was a job whose `next_run_at` still pointed at the fire already in flight.

**Mechanism as I read it:** a worker process spawned for job A runs the tick for *all* jobs. Its in-flight guard (`_running_job_ids` + `_running_since`) is per-process, so a job in flight in another process still looks due — `next_run_at` is not advanced at dispatch — and it gets re-fired while inside its catch-up window. That re-fire blocks on the per-job fire fence for 30s, fails closed, and the collapsed fail-closed path makes the *other* process's healthy run kill itself. So this is reachable on any gateway where jobs share a tick or where a short-period watchdog worker starts while a longer job is running — not a rare race.

I can't tell from the logs whether the claim record is actually overwritten by the failed acquire or whether the running worker's sampled flag flips on its own re-read of the fence; if it's the latter, the failed acquire is destructive on its own and worth asserting in a test.

**Side note for other reporters:** the work is not lost — the digest was delivered both times; only the status/dedup bookkeeping is wrong (`delivered` + `failed`, and the recovery text tells the operator the result was discarded when it wasn't). Same family as #105861 / #105882.

**Mitigation on our side (no code change):** moving data collection into a pre-run `script` cut the job from ~11 min to ~2m53s, which shrinks the window in which a re-fire can land mid-run. That's damage control, not a fix.

kshitijk4poor:
The fix for this bug landed on main via https://github.com/NousResearch/hermes-agent/pull/109310 (merge 9a60a7f3167e): `heartbeat_fire_claim` now refreshes under `_jobs_lock` only and no longer waits on the per-thread fire fence its own run holds, so a delivery/agent turn longer than the 30s fence timeout is no longer misread as ownership loss. `_refresh_claim` still CASes on `claim["by"]`, so a real takeover keeps returning `False` (pinned by test).

Thanks — the tri-state/unverifiable classification was a correct reading of the failure. The landed fix removes the fence wait from the heartbeat entirely, so `False` now means only "another owner holds the claim"; the grace-path routing this PR adds becomes unnecessary. Re @jrodriguezruibal's production confirmation: that routine "could not verify" path is exactly what is gone. Closing with credit.

PR #97566, opened 2026-08-29
fix(chronos): use hosted agent credential

### LINKED ISSUE #97494: [Bug]: Chronos provisioning rejects hosted Nous Cloud agents with HTTP 403
## Bug Description

The Chronos cron provider cannot provision scheduled jobs for a Hermes Agent running as a hosted Nous Cloud agent. The job is stored locally, but registration with the Nous agent-cron service fails with HTTP 403.

## Steps to Reproduce

1. Run Hermes Agent as a hosted Nous Cloud agent.
2. Configure the cron provider as `chronos`.
3. Create a recurring cron job from Hermes Desktop or the dashboard.
4. Observe the gateway logs during Chronos registration.

## Expected Behavior

A hosted Nous Cloud agent should be recognized as a provisioned agent and Chronos should register the scheduled trigger successfully.

## Actual Behavior

The job is persisted, but Chronos cannot arm or re-arm it. The gateway logs:

```text
POST /api/agent-cron/provision returned 403:
{"error":"invalid_client","error_description":"agent-cron is only available to a provisioned agent (agent:* client or bootstrap session)"}
```

This occurs when initially arming jobs and again when re-arming recurring jobs after execution.

## Environment

- Deployment: Hosted Nous Cloud Hermes Agent
- Hermes version: v0.20.6
- Client: Hermes Desktop and dashboard
- Cron provider: Chronos
- Inference provider: Nous Portal

## Error Output

```text
WARNING cron.chronos: Chronos failed to arm job: POST /api/agent-cron/provision returned 403: {"error":"invalid_client","error_description":"agent-cron is only available to a provisioned agent (agent:* client or bootstrap session)"}
```

## Additional Context

The main agent can authenticate to Nous inference and execute the cron job successfully. The failure is specific to the Chronos provisioning endpoint and appears to indicate that the hosted agent's client identity or provisioning state is not accepted by the agent-cron service.

The local cron scheduler continues to execute jobs while the gateway is running, but external scheduling and re-arming are unavailable.

Related issues:

- #62951, bundled Chronos plugin loading failure
- #69715, profile-specific Chronos fire-token validation
- #97475, agent-owned delivery bridge for spawned cron jobs

Please avoid including agent IDs, callback URLs, email addresses, API keys, or other deployment-specific identifiers in this issue.

wesleysimplicio:
Closes #97494

## Summary
- accept persisted hosted `agent_key` credentials when deciding Chronos availability
- resolve the same runtime agent credential used by Nous inference for NAS calls
- add regression coverage for hosted bootstrap credentials

Chronos provisioning must authenticate with the gateway's agent-scoped bootstrap/invoke credential. Using the generic session token can be rejected by `/api/agent-cron/provision` even while inference continues to work.

## Validation
- `python -m pytest -q tests/plugins/test_chronos_cron.py` (15 passed)
- `git diff --check`
- the existing JWT verifier tests remain dependency-blocked here because PyJWT is not installed.

### DIFF
--- a/plugins/cron_providers/chronos/__init__.py
+++ b/plugins/cron_providers/chronos/__init__.py
@@ -83,7 +83,10 @@ class ChronosCronScheduler(CronScheduler):
         try:
             from hermes_cli.auth import get_provider_auth_state
             state = get_provider_auth_state("nous") or {}
-            return bool(state.get("access_token"))
+            # Hosted installs may persist the bootstrap/invoke credential as
+            # ``agent_key`` while the refreshable user token is absent. Both
+            # are valid signals for the lazy runtime resolver below.
+            return bool(state.get("access_token") or state.get("agent_key"))
         except Exception:
             return False
 
--- a/plugins/cron_providers/chronos/_nas_client.py
+++ b/plugins/cron_providers/chronos/_nas_client.py
@@ -41,9 +41,22 @@ class NasCronClient:
     # -- auth -------------------------------------------------------------
 
     def _access_token(self) -> str:
-        """The agent's existing Nous Portal access token (refresh-aware)."""
-        from hermes_cli.auth import resolve_nous_access_token
-        return resolve_nous_access_token()
+        """Resolve the agent-scoped Nous credential used by managed services.
+
+        ``resolve_nous_access_token`` exposes the refreshable user session,
+        while hosted gateways may need the runtime ``agent_key`` (the
+        bootstrap/invoke JWT seeded for that gateway). The runtime resolver
+        refreshes and selects that credential consistently with inference, and
+        prevents Chronos from presenting a generic user token to
+        ``agent-cron`` (#97494).
+        """
+        from hermes_cli.auth import resolve_nous_runtime_credentials
+
+        credentials = resolve_nous_runtime_credentials()
+        token = credentials.get("api_key")
+        if not isinstance(token, str) or not token.strip():
+            raise NasCronClientError("Nous runtime credentials did not contain an access token")
+        return token.strip()
 
     def _headers(self) -> Dict[str, str]:
         return {
--- a/tests/plugins/test_chronos_cron.py
+++ b/tests/plugins/test_chronos_cron.py
@@ -61,6 +61,34 @@ def test_is_available_false_without_config(temp_home, monkeypatch):
     assert ChronosCronScheduler().is_available() is False
 
 
+def test_is_available_accepts_hosted_agent_key(temp_home, monkeypatch):
+    from plugins.cron_providers.chronos import ChronosCronScheduler
+
+    monkeypatch.setattr(
+        "plugins.cron_providers.chronos._cfg",
+        lambda *k, default="": "https://agent.example/"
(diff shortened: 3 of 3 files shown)


Enough1122:

**Overall:** Correctly switches Chronos's availability probe and NAS client credential resolution to the hosted/agent-scoped credential path (#97494).

**What it does**
- `plugins/cron_providers/chronos/__init__.py:_have_nous_token` now treats `agent_key` (bootstrap/invoke JWT persisted on hosted installs) as a valid availability signal alongside the refreshable `access_token`.
- `plugins/cron_providers/chronos/_nas_client.py:_access_token` resolves through `hermes_cli.auth.resolve_nous_runtime_credentials()` (the same resolver used by inference) and selects `credentials["api_key"]`, falling back to a loud `NasCronClientError` when the token is missing/blank — so Chronos no longer presents a generic user token to `agent-cron`.
- Tests cover the hosted-`agent_key` availability path and the runtime-credential token resolution.

**Non-blocking notes**
- The two sources are different: `_have_nous_token` reads `get_provider_auth_state("nous").get("agent_key")`, while `_access_token` reads `resolve_nous_runtime_credentials().get("api_key")`. The tests mock both consistently, so a mismatch between the two stores (e.g. `agent_key` present in auth state but surfaced under a different key — or not at all — by the runtime resolver) would make `is_available()` return True while `_access_token()` raises. Worth a test that mocks the two stores divergently to pin that they agree, or a single-source-of-truth lookup.
- `is_available()` returning True before `_access_token()` raises means `NasCronClient` could fail at first use rather than being cleanly reported as unavailable — acceptable, but a debug log at `_access_token` failure would aid diagnosis.

No blocking issues; the change is small, well-scoped, and correctly targeted at the hosted-credential gap.

teknium1:
Thanks @wesleysimplicio for digging into #97494 — the direction (Chronos must present the agent's provisioned identity) is right, but this diff can't change what reaches NAS on current `main`.

**Why it's a no-op on the wire:** `hermes_cli/auth_nous.py::_set_nous_agent_key_from_invoke_jwt` sets `agent_key = access_token`, so `resolve_nous_runtime_credentials()["api_key"]` and `resolve_nous_access_token()` return the same bearer. E2E with a hosted-shaped `auth.json` (client `hermes-cli-vps`): both resolvers and `NasCronClient._headers()` yield the identical token. The `is_available()` change is likewise inert (`agent_key` never exists without `access_token`).

**What actually 403s:** NAS (hermes-portal `server/agent-cron/instance-auth.ts`) accepts an `agent:*` client or a `hermes-cli-vps` bootstrap session whose `sid` matches `AgentInstance.bootstrapSessionId`. Containers hitting `invalid_client` hold a plain `hermes-cli` user login in `auth.json` (typically after a `hermes auth logout nous` + re-login, which *replaces* the bootstrap session). That is fixed NAS-side by re-minting the bootstrap session, not by any token path in Hermes.

Superseded by #115573, which makes Chronos treat `403 invalid_client` as the deterministic identity rejection it is: one warning naming the real remedy, NAS left alone, built-in ticker takes over so jobs fire on time. Closing this one; credited in that PR.

Issue #54465, opened 2026-06-28
Preflight context compression can wedge resumed sessions when auxiliary compression times out


rodboev:
### Summary

A large CLI session can become effectively unresumable when preflight context compression is required and the auxiliary compression endpoint times out. Every new user message enters synchronous preflight compression before the assistant can respond. If the auxiliary route is slow/hung, the send appears stalled for many minutes. Killing the CLI and resuming the session repeats the same compression attempt because the failure cooldown is in-memory only.

There is also a lock-safety issue: the SQLite compression lock TTL defaults to 300s, but a single auxiliary compression call can exceed that. In the incident below, the first timeout surfaced after about 361s, so the lock expired while the compressor was still alive.

### Impact

This presents to the user as: "the session will not continue or compact; every message stalls." The persisted messages do not appear corrupt, but the session is practically wedged because it must compact before any next model call and the compaction path cannot fail fast.

### Real session evidence

Session: `20260628_155140_021192`
Surface: CLI
Main model/provider: `gpt-5.5` via `openai-codex`
Local checkout observed during analysis: `d3d621f7c` on `main`
Runtime packages observed: `openai 2.14.0`, `httpx 0.28.1`, `httpcore 1.0.9`

Session DB shape from `state.db`:

- source: `cli`
- model: `gpt-5.5`
- end_reason: `cli_close`
- session `message_count`: `308`
- active DB messages: `308`
- all DB messages: `498`
- compacted old rows: `190`
- active role counts: assistant `124`, tool `171`, user `13`
- active content chars: `614,776`
- largest active message: `49,170` chars (`skill_view` output)
- active context contained an earlier compaction handoff summary: message id `6056`, `23,116` chars
- last active user message: message id `6358`

Token estimates observed in logs:

- A turn hit compression around `231,496` input tokens.
- The next user message immediately hit preflight compression: `~307,285 tokens >= 231,200 threshold (model gpt-5.5, ctx 272,000)`.
- A local rough estimate from active DB messages alone was `195,148` tokens; request-time estimate is higher because system prompt and tools are included.

Relevant `agent.log` excerpt from the first compression attempt:

```text
2026-06-28 16:46:30,935 INFO [20260628_155140_021192] agent.model_metadata: Could not detect context length for model 'gpt-5.5' at http://127.0.0.1:8788/v1/ — defaulting to 256,000 tokens (probe-down). Set model.context_length in config.yaml to override.
[...]

Issue #51513, opened 2026-06-23
fix(vision): fallback chain completely non-functional — 5 bugs, complete fix with verification


saved-j:
### The Problem

Gemini is the go-to free vision provider — great quality, zero cost. But it's unreliable:

- **Rate limits / quota exhaustion** — free tier hits daily caps
- **Geo-restrictions** — users behind VPNs (China, corporate networks) get `FAILED_PRECONDITION` when the VPN drops
- **Service outages** — 503 "high demand" during peak hours

When Gemini fails, vision should fall back to the next provider in the chain. **It doesn't.** Five independent bugs compound to make `auxiliary.vision.fallback_chain` completely non-functional.

### Why Vision Needs Its Own Fallback Path

Other auxiliary tasks (compression, title generation, etc.) are text-only — any text model works as fallback. Vision is different: **it requires a vision-capable model**. Sending an image to a text-only model returns garbage or 404s.

This means vision needs its own dedicated fallback chain with vision-capable models, separate from the generic auxiliary fallback.

**Example chain:**
```
Gemini 2.5 Flash (free) → Xiaomi mimo-v2.5 (paid) → ollama/holo-3.1:4b (local)
```

---

### Bug 1: `supports is None → return True` (line 4705)

**File:** `agent/auxiliary_client.py:4705`

When the model database has no capability info for a model, `_main_model_supports_vision()` assumes it **supports** vision. This sends images to the text-only main model instead of routing through the aux chain.

**Before (vanilla):**
```python
    if supports is None:
        # No capability data — keep current behaviour and let the call attempt
        # happen rather than silently skipping. This avoids false-positive
        # skips for new/custom providers.
        return True
```

**Fix (config workaround — add `supports_vision: false` per-model):**
```yaml
providers:
  Xiaomi-TP:
    models:
      mimo-v2.5-pro:
        supports_vision: false    # ← text-only, forces aux chain
      mimo-v2.5:
        supports_vision: true     # ← vision-capable
```

**Upstream code fix (pending):**
```python
    if supports is None:
        return False  # unknown → assume text-only → use aux chain
```

---

### Bug 2: `await` on sync OpenAI client (line 1007)

**File:** `tools/vision_tools.py:1007`

`_try_configured_fallback_chain()` returns a **sync** `OpenAI()` client, but the code uses `await` on it. This raises `TypeError` which is masked by `raise _api_err from _fb_err`.

**Before (vanilla):**
```python
[...]

PR #87087, opened 2026-08-15
fix(vision): allow configured timeout fast-fail

mybryan-sketch:
## Summary

- add `auxiliary.vision.retry_on_timeout` with backward-compatible default
  behavior
- when set to `false`, skip the second same-provider attempt after a vision
  timeout
- continue into the configured fallback chain without changing non-timeout
  retry behavior

## Why

Document and image-heavy workloads can spend a second full timeout budget
retrying the same provider before a healthy fallback is attempted.

## Current-main verification

- rebased onto `73f68362b` (2026-09-02)
- focused fast-fail regression: passed
- full auxiliary-client module: 188 passed; one unrelated Anthropic OAuth
  refresh test also fails on unmodified current main
- Ruff and `git diff --check`: passed

PR #97572, opened 2026-08-29
fix(auxiliary): skip same-provider retry on a vision full-budget timeout

xmhuangzhijun-hue:
## What does this PR do?

Issue #54465 established that a same-provider retry after a **full-budget timeout** costs a second whole `timeout` window before the fallback chain is reached — doubling the user-visible stall — and that compression must not pay it because it sits on a critical path.

The guard added for that is spelled `task == "compression"`, at both retry sites:

```python
if task == "compression" and _is_timeout_error(transient_err):
```

So **vision still retries**, and vision sits on the interactive path. The cost is identical and the stall is more visible: the turn holding the image cannot answer, and because turns are serialised the following user messages queue behind it. Two sequential full-budget timeouts against an unhealthy vision provider is a long stall for something the fallback chain could have served immediately.

This replaces the string comparison at both sites (sync `call_llm` and `async_call_llm`) with a shared `_TIMEOUT_NO_RETRY_TASKS = frozenset({"compression", "vision"})`, so the two paths cannot drift apart again.

Behaviour is unchanged for every other task. Fast blips (a streaming-close or a 5xx) still retry — only full-budget timeouts on those two tasks skip straight to fallback.

Found while running a deployment with a configured `auxiliary.vision` provider on a link that times out intermittently.

## Related Issue

Precedent and rationale: #54465

Not a duplicate of #51513 — that fixes five separate defects **inside** the vision fallback chain (capability detection, sync/async client misuse, geo-block and `RemoteProtocolError` classification, chain iteration). This is about the wasted timeout window **before** that chain is reached, and the two changes are independent.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `agent/auxiliary_client.py` — added `_TIMEOUT_NO_RETRY_TASKS`; both retry sites now test membership instead of comparing to the literal `"compression"`; comments updated to state why vision qualifies. Log lines carry the task name rather than hardcoding it.
- `tests/agent/test_auxiliary_client.py` — two tests in `TestTransientTransportRetry`:
  - `test_vision_skips_same_provider_retry_on_timeout` — primary is tried exactly once, then fallback.
  - `test_non_critical_task_still_retries_same_provider_on_timeout` — an ordinary task keeps its one same-provider retry, proving the change is scoped rather than blanket.

## Testing

The vision test has teeth: reverting the source change fails it while leaving the scoping test green.

`pytest tests/agent/test_auxiliary_client.py`:

- without this change: **5 failed, 180 passed**
- with this change: **5 failed, 182 passed** (the +2 are the new tests)

The 5 pre-existing failures are unrelated to this path and reproduce on a clean checkout of `main`.

One note for review: vision resolves its client through `resolve_vision_provider_client()` rather than `_get_cached_client()`, so the new test patches that instead — the retry block under test is shared by both paths.

### DIFF
--- a/agent/auxiliary_client.py
+++ b/agent/auxiliary_client.py
@@ -4697,6 +4697,14 @@ def _is_invalid_aux_response_error(exc: Exception) -> bool:
     )
 
 
+# Auxiliary tasks that sit on a user-visible critical path. A same-provider
+# retry after a full-budget timeout costs another whole ``timeout`` window
+# before the fallback chain is reached, so these skip it and fall through
+# immediately. Fast blips (a streaming-close or a 5xx) still retry, since
+# those are cheap. See issue #54465 for the compression case.
+_TIMEOUT_NO_RETRY_TASKS = frozenset({"compression", "vision"})
+
+
 def _evict_cached_clients(provider: str) -> None:
     """Drop cached auxiliary clients for a provider so fresh creds are used."""
     normalized = _normalize_aux_provider(provider)
@@ -9978,18 +9986,21 @@ def _call_llm_impl(
             if not _is_transient_transport_error(transient_err):
                 raise
             # Compression is on the critical preflight path: a user cannot
-            # continue or resume an oversized session until it compacts. A
-            # same-provider retry on a timeout means another full ``timeout``-
-            # long wall-clock block before the except-chain below can fall
-            # back — doubling the user-visible stall (issue #54465). Skip the
-            # same-provider retry for compression on a full-budget timeout and
+            # continue or resume an oversized session until it compacts.
+            # Vision is on the interactive path: the turn holding the image
+            # cannot answer, and because turns are serialised the following
+            # user messages stall behind it. In both cases a same-provider
+            # retry on a timeout means another full ``timeout``-long
+            # wall-clock block before the except-chain below can fall back —
+            # doubling the user-visible stall (issue #54465). Skip the
+            # same-provider retry for those tasks on a full-budget timeout and
             # fall straight through to provider/model fallback; fast blips (a
             # streaming-close or a 5xx) still retry, since those are cheap.
-            if task == "compression" and _is_timeout_error(transient_err):
+            if task in _TIMEOUT_NO_RETRY_TASKS and _is_timeout_error(transient_err):
                 logger.info(
-                    "Auxiliary compression: timeout on the critical path; "
+                    "Auxiliary %s: timeout on the critical path; "
                     "skipping same-provider retry and falling back: %s",
-                    transient_err,
+                    task, transient_err,
                 )
                 raise
             _max_transient_retries = _transient_retry_count()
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Sound extension of the "skip same-provider retry on full-budget timeout" policy from compression to vision — both sit on user-visible critical paths where a second full `timeout` window stalls the turn.

**What it does**
- Introduces a module-level `_TIMEOUT_NO_RETRY_TASKS = frozenset({"compression", "vision"})` and uses `task in _TIMEOUT_NO_RETRY_TASKS and _is_timeout_error(...)` in both the sync (`_call_llm_impl`) and async (`_acreate`) paths, replacing the previous `task == "compression"` checks.
- Log lines updated to include the task name, keeping the sync/async messages consistent.
- Tests add a vision-timeout fallback case (asserts exactly one primary attempt + fallback) and a negative control proving non-critical tasks (`title`) still retry the same provider once — confirming this is not a blanket behavior change.

**Non-blocking notes**
- The frozenset is keyed on the exact task string `"vision"`. If any call site passes a variant (e.g. `"image"`, `"vision_analysis"`) it silently misses the skip. Worth a quick grep that every vision-path call uses the literal `task="vision"`; a unit test that exercises the actual call-site constant would pin it.
- The `compression` vs `vision` critical-path reasoning is documented well; the comment references the same issue #54465 for both, which is accurate for compression but the vision rationale is broader (serialised turn stall) — the inline comment covers this correctly.

No blocking issues; the change is minimal, symmetric across sync/async, and well-tested.

xmhuangzhijun-hue:
Confirmed against the PR tree: the production vision entry points use the exact `"vision"` task value. `tools/vision_tools.py` builds `call_kwargs` with `"task": "vision"` for both image and video analysis, and `tools/browser_camofox.py` passes `task="vision"`. I found no production call site using an `image` or `vision_analysis` variant. The test also reaches `call_llm(task="vision")`, so the frozenset key matches the real path.

kshitijk4poor:
Triage note (perf/P2 sweep, verified against `origin/main` `30b83ab7b1`):

**Canonical for the vision-timeout pair (#97572 / #87087).** Premise live on main: only `task == "compression"` skips the same-provider retry, at both the sync (`~10632`) and async (`~11465`) sites; `tools/vision_tools.py` drives `async_call_llm(task="vision")`.

Your diff no longer applies cleanly because f50b5bb0fa added a carve-out on the sync site after your base (a no-progress timeout — dead stream, zero output within 60 s — is cheap and still retries; only stalls/hard-ceiling skip). Salvaged onto current main as #101570 with your commit cherry-picked (authorship preserved; conflict resolved keeping that carve-out) plus two follow-ups of ours: mirror the carve-out on the async site (it never had it — a stillborn async vision stream would have skipped to fallback where sync retries), and fold the three-clause decision into one `_should_skip_same_provider_retry(task, exc)` predicate next to `_TIMEOUT_NO_RETRY_TASKS` so the two sites can't drift again.

Benchmark (macOS / Apple Silicon; stubbed network — primary sleeps `TIMEOUT_S=0.5` then raises an `APITimeoutError` look-alike, fallback answers instantly; `task="vision"`, ×5, median):

| | sync | async | primary attempts |
|---|---|---|---|
| main | 1015 ms | 1010 ms | 2 |
| #101570 | 517 ms | 511 ms | 1 |

What actually improved: one timeout window instead of two before fallback runs — with the real vision timeout that is on the order of minutes saved per timed-out image, on the path that blocks the next user message. Tests: your two tests + `test_codex_aux_no_progress_timeout.py` 15/15 on the salvage; reverting the change fails the vision-skip tests.

kshitijk4poor:
Merged via #101570 (rebase-merge, your commit cherry-picked with authorship preserved — merge commit `c4e394cdf8` on main). Thanks @xmhuangzhijun-hue.

On the review notes here: the literal `"vision"` task key you confirmed is also what I verified on current main (`tools/vision_tools.py`, `tools/browser_camofox.py`). The sync/async "same reasoning" comment concern is resolved structurally — both sites now call one `_should_skip_same_provider_retry(task, exc)` predicate that carries the rationale (and the no-progress carve-out from f50b5bb0fa) in a single place. Closing this PR since the change is on main.

PR #97574, opened 2026-08-29
fix(desktop): preserve model submenu pointer intent

### LINKED ISSUE #97505: [Bug]: Desktop model submenu closes during diagonal pointer travel
## Bug description

In Hermes Desktop's model picker, moving the pointer diagonally from a model row toward a lower item in its portaled Thinking / Effort / Fast submenu can cross a sibling row. That sibling's hover handler immediately replaces the active submenu, so the intended submenu disappears before the pointer reaches the lower option.

This is an observed pointer trajectory failure, not screenshot-only geometry. It is distinct from #86966 (affordance/keyboard access), #80798 (stacking/clickability), and #55856 / #55858 (scroll-position drift).

## Steps to reproduce

1. Open the Desktop composer model picker.
2. Hover a model row to open its options submenu.
3. Move diagonally toward a lower effort option, crossing the neighboring model-row region.
4. The neighboring row opens and replaces the intended submenu before the destination can be selected.

## Expected behavior

The shortest natural diagonal path into the already-open submenu should remain usable. A deliberate vertical move to another model row should still switch promptly.

## Integration detail

Radix DropdownMenu already provides bounded pointer grace inside its nested-menu primitive. The Hermes model catalog also opens sibling submenus directly from row-level pointer movement, which can replace the active submenu before Radix's portaled content receives pointer entry. The fix should integrate with that existing grace rather than add a second global menu engine or an invisible overlay.

## Acceptance criteria

- A real Electron pointer trajectory reaches and selects the lowest effort option through the sibling-row region.
- Deliberate vertical movement switches to the sibling submenu promptly.
- Diagonal movement followed by Escape closes cleanly, restores composer focus, and reopens without stale intent.
- The behavior is pointer-only: keyboard, touch/click, Enter/Space, Escape, outside-click, focus restoration, and screen-reader semantics remain unchanged.
- No global pointer listener or timer survives menu close; pointer history remains small and bounded.
- Portaled and scrollable menu geometry is exercised in the runtime test.

## Environment

Hermes Desktop model catalog (`ModelCatalogMenu` / `ModelEditSubmenu`) using Radix DropdownMenu with portaled submenu content.

wesleysimplicio:
Closes #97505

## Summary
- retain the active model edit submenu when a pointer crosses a sibling row along a natural diagonal path toward the portaled content
- let deliberate vertical movement switch rows immediately
- use a bounded last-pointer sample and Radix's existing submenu behavior, with no global pointer listener or persistent timer
- add focused geometry regression coverage

## Validation
- `git diff --check`
- desktop Vitest could not start in this environment because the installed dependency tree does not provide `vitest/config`

The guard only prevents the sibling trigger's pointermove default when the projected path still intersects the open submenu (with a small edge tolerance). Once the pointer enters the submenu, or the path turns away, Radix resumes normal submenu handling and Escape/focus behavior remains owned by the primitive.

### DIFF
--- a/apps/desktop/src/app/shell/model-catalog-menu.test.tsx
+++ b/apps/desktop/src/app/shell/model-catalog-menu.test.tsx
@@ -11,7 +11,7 @@ import {
   setVisibleModels
 } from '@/store/model-visibility'
 
-import { ModelCatalogMenu, type ModelMenuController } from './model-catalog-menu'
+import { isPointerMovingTowardSubmenu, ModelCatalogMenu, type ModelMenuController } from './model-catalog-menu'
 
 // Radix calls these on open; jsdom doesn't implement them.
 beforeAll(() => {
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: #97508 also fixes #97505's diagonal submenu-travel failure. This PR uses a bounded last-pointer projection, while #97508 uses a different pointer-grace implementation; these are competing approaches rather than duplicates.

PR #97576, opened 2026-08-29
feat(image-gen): add direct Gemini backend

### LINKED ISSUE #97474: Feature: Native Google AI Studio (Gemini/Nano Banana) image-generation backend using GOOGLE_API_KEY directly
## Feature Description

Add a native Google AI Studio image-generation backend to `image_gen` that talks directly to `generativelanguage.googleapis.com` using a plain `GOOGLE_API_KEY` / `GEMINI_API_KEY` — no FAL.ai / OpenRouter / Nous Portal account required.

## Motivation

Users who only have a Google AI Studio API key (no FAL or OpenRouter account) currently cannot generate images with Gemini/"Nano Banana" image models through Hermes at all. `image_gen` supports FAL, OpenAI, OpenAI-Codex, xAI, Krea, DeepInfra, and OpenRouter (which can proxy `google/gemini-3-pro-image`), but none of these read `GOOGLE_API_KEY`/`GEMINI_API_KEY` or call Google's API directly.

Separately, Hermes already has a native "Gemini" **text/chat** provider (`plugins/model-providers/gemini/__init__.py`) that reads `GOOGLE_API_KEY`/`GEMINI_API_KEY` and talks to `https://generativelanguage.googleapis.com/v1beta` — so the credential-resolution plumbing for a direct Google integration already exists in the codebase, it's just never wired up for image generation.

This creates an inconsistent story: a user can use their Google key for chat, but is forced to sign up with a second, unrelated vendor (FAL or OpenRouter) just to get image generation with Google's own model.

## Proposed Solution

Add a new `image_gen` plugin, e.g. `plugins/image_gen/gemini/__init__.py`, modeled closely on the existing `plugins/image_gen/openrouter/__init__.py` (`OpenRouterCompatImageProvider`) as reference for how an `ImageGenProvider` subclass should be structured (config resolution order, `save_b64_image`/`save_url_image` helpers, `success_response`/`error_response` contract, image-to-image reference handling).

Differences vs. the OpenRouter plugin:
- Auth: resolve `GOOGLE_API_KEY` / `GEMINI_API_KEY` directly (same env vars already used by `plugins/model-providers/gemini/__init__.py`), not an OpenRouter/Nous runtime resolution.
- Endpoint: call Google's native `generateContent` REST endpoint directly, e.g.
  `POST https://generativelanguage.googleapis.com/v1beta/models/{model}:generateContent?key=$GOOGLE_API_KEY`
  with `generationConfig.responseModalities: ["IMAGE"]` (per Gemini image-generation docs), rather than the OpenAI-style `/chat/completions` + `modalities` shape.
- Default model(s): `gemini-3-pro-image-preview` / `gemini-2.5-flash-image` ("Nano Banana" / "Nano Banana Pro") — whichever Google currently ships as GA vs preview.
- Image extraction: Google returns generated images as inline base64 data in `candidates[].content.parts[].inlineData.data`, different from OpenRouter's `choices[0].message.images[].image_url.url` shape.
- Reference images (image-to-image / editing) should map to Gemini's multi-part `contents` with `inlineData` image parts, same underlying capability the OpenRouter path already exposes for this model family.

This should register as its own backend choice in `hermes tools` → Image Generation (e.g. "Google AI Studio (direct)"), alongside the existing FAL/OpenAI/xAI/Krea/DeepInfra/OpenRouter options, with its own `image_gen.gemini.model` config scope following the same precedence pattern (`explicit call override → GEMINI_IMAGE_MODEL env var → image_gen.gemini.model → image_gen.model → default`).

## Alternatives Considered

- **Use OpenRouter as a proxy for `google/gemini-3-pro-image`.** Works today, but requires a separate OpenRouter (or Nous Portal) account/key — defeats the purpose for a user who specifically only wants to use their existing Google API key.
- **Use FAL.ai's `fal-ai/nano-banana-pro` model.** Also works today, same drawback — requires a FAL account/key instead of the Google key the user already has.
- **Do nothing / document the OpenRouter workaround.** Leaves the actual gap (no first-party Google image path) unaddressed; the text-provider precedent for direct Google auth already exists in the codebase, so a native path is a natural, incremental extension rather than new infrastructure.

## Related code pointers

- `plugins/model-providers/gemini/__init__.py` — existing native Gemini **text** provider; shows the `GOOGLE_API_KEY`/`GEMINI_API_KEY` resolution and base URL already used elsewhere in the codebase.
- `plugins/image_gen/openrouter/__init__.py` — closest existing `ImageGenProvider` implementation to use as a structural template (config precedence, response saving helpers, image-to-image reference handling).
- `agent/image_gen_provider.py` — the `ImageGenProvider` ABC / shared helpers (`save_b64_image`, `save_url_image`, `success_response`, `error_response`) any new backend must implement against.
- `website/docs/user-guide/features/image-generation.md` — user-facing docs listing all currently supported backends/models; would need a new row/section for this backend.

wesleysimplicio:
Closes #97474

## Summary

- add a native Google AI Studio image-generation provider using GOOGLE_API_KEY or GEMINI_API_KEY
- call Gemini generateContent directly with the x-goog-api-key header instead of placing credentials in the request URL
- request both TEXT and IMAGE modalities so safety blocks and text-only responses remain diagnosable
- expose the current stable Gemini image-model catalog, including Gemini 3.1 Flash Image as the default
- enforce the legacy Gemini 2.5 reference-image limit while retaining the 14-image capability for newer Gemini image models
- support URL/data/local reference images through inlineData and cache generated base64 output under HERMES_HOME
- expose the backend and model catalog through the existing plugin/picker registry and document setup

## Validation

| Command | Result |
|---|---|
| `python -m py_compile plugins/image_gen/gemini/__init__.py tests/plugins/image_gen/test_gemini_provider.py` | Passed |
| `uv run --extra dev pytest -q tests/plugins/image_gen/test_gemini_provider.py tests/agent/test_image_gen_registry.py` | 12 passed |
| `uv run --extra dev pytest -q tests/tools/test_image_generate_schema.py` | 9 passed, 29 subtests passed |
| `uv run ruff check plugins/image_gen/gemini/__init__.py tests/plugins/image_gen/test_gemini_provider.py` | All checks passed |
| `git diff --check` | Passed |
| Fail-before regression run against the pre-fix implementation | 4 failed, 4 passed |
| Pass-after regression run | 12 passed |

The fail-before run reproduced the missing header assertion, the legacy reference-limit failure, the missing diagnostic text, and the stale catalog assertion. The pass-after run covers those cases, the registry integration, and the schema contract.

Desktop/editor and full dashboard suites were not part of this backend change; the local environment lacks some optional frontend dependencies (notably `vitest/config` and `fastapi`).

### DIFF
--- a/plugins/image_gen/gemini/__init__.py
+++ b/plugins/image_gen/gemini/__init__.py
@@ -281,7 +281,7 @@ class GeminiImageGenProvider(ImageGenProvider):
             response = requests.post(
                 endpoint,
                 params={"key": api_key},
-                headers={"Content-Type": "application/json"},
+                headers={"x-goog-api-key": api_key, "Content-Type": "application/json"},
                 json=payload,
                 timeout=timeout,
             )
--- a/plugins/image_gen/gemini/__init__.py
+++ b/plugins/image_gen/gemini/__init__.py
@@ -40,16 +40,29 @@ _ASPECT_RATIOS = {
     "portrait": "9:16",
 }
 _MODELS: Dict[str, Dict[str, str]] = {
-    "gemini-3-pro-image-preview": {
+    "gemini-3.1-flash-lite-image": {
+        "display": "Nano Banana 2 Lite (Gemini 3.1 Flash Lite Image)",
+        "speed": "fast",
+        "strengths": "Lowest latency and cost; efficient image generation and editing",
+    },
+    "gemini-3.1-flash-image": {
+        "display": "Nano Banana 2 (Gemini 3.1 Flash Image)",
+        "speed": "balanced",
+        "strengths": "General-purpose generation, 4K output, and multi-reference editing",
+    },
+    "gemini-3-pro-image": {
         "display": "Nano Banana Pro (Gemini 3 Pro Image)",
-        "speed": "~30s",
-        "strengths": "Highest fidelity, prompt adherence, and image editing",
+        "speed": "premium",
+        "strengths": "Highest fidelity, reasoning, 4K output, and search grounding",
     },
     "gemini-2.5-flash-image": {
         "display": "Nano Banana (Gemini 2.5 Flash Image)",
-        "speed": "~10s",
-        "strengths": "Fast generation and image editing",
+        "speed": "fast",
+        "strengths": "Legacy fast generation and image editing",
     },
+}
+_MODEL_REFERENCE_LIMITS = {
+    "gemini-2.5-flash-image": 3,
 }
 
 
@@ -91,7 +104,9 @@ def _resolve_model(explicit: Optional[str] = None) -> str:
 def _api_key() -> Optional[str]:
     """Return the direct Google key, keeping the text-provider precedence."""
     try:
-        return (get_secret("GOOGLE_API_KEY") or get_secret("GEMINI_API_KEY") or "").strip() or None
+        return (
+            get_secret("GOOGLE_API_KEY") or get_secret("GEMINI_API_KEY") or ""
+        ).strip() or None
     except Exception as exc:  # noqa: BLE001 - scoped-secret failures are auth failures
         logger.debug("Google AI Studio credential lookup failed: %s", exc)
         return None
@@ -107,8 +122,12 @@ def _load_reference_image(reference: str) -> Tuple[bytes, str]:
 
         response = requests.get(reference, timeout=60)
         response.raise_for_status()
-        content_type = (response.headers.get("Content-Type") or "").split(";", 1)[0].strip()
-        return response.content, content_type if content_type.startswith("image/") else "image/png"
+        content_type = (
+            (response.headers.get("Content-Type") or "").split(";", 1)[0].strip()
+        )
+        return response.content, content_type if content_type.startswith(
+            "image/"
+        ) else "image/png"
 
     if lower.startswith("data:"):
         header, separator, encoded = reference.partition(",")
@@ -233,7 +252,8 @@ class GeminiImageGenProvider(ImageGenProvider):
         if isinstance(image_url, str) and image_url.strip():
             sources.append(image_url.strip())
         sources.extend(normalize_reference_images(reference_image_urls) or [])
-        sources = sources[:_MAX_REFERENCE_IMAGES]
+        reference_limit = _MODEL_REFERENCE_LIMITS.get(model, _MAX_REFERENCE_IMAGES)
+        sources = sources[:reference_limit]
 
         parts: List[Dict[str, Any]] = [{"text": prompt}]
         for source in sources:
@@ -248,32 +268,25 @@ class GeminiImageGenProvider(ImageGenProvider):
                     prompt=prompt,
                     aspect_ratio=aspect,
                 )
-            parts.append(
-                {
-                    "inlineData": {
-                        "mimeType": mime,
-                        "data": base64.b64encode(data).decode("ascii"),
-                    }
+            parts.append({
+                "inlineData": {
+                    "mimeType": mime,
+                    "data": base64.b64encode(data).decode("ascii"),
                 }
-            )
+            })
 
         payload = {
             "contents": [{"role": "user", "parts": parts}],
             "generationConfig": {
                 "responseModalities": ["TEXT", "IMAGE"],
                 "imageConfig": {"aspectRatio": _ASPECT_RATIOS[aspect]},
-                "imageConfig": {"aspectRatio": _ASPECT_RATIOS[aspect]},
             },
         }
-        timeout = kwargs.get("timeout", 120)
         timeout = kwargs.get("timeout", 180)
         try:
             timeout = max(1.0, float(timeout))
         except (TypeError, ValueError):
             timeout = 180
-            timeout = max(1.0, float(timeout))
-        except (TypeError, ValueError):
-            timeout = 120
 
         endpoint = f"{BASE_URL}/models/{quote(model, safe='')}:generateContent"
         try:
@@ -281,7 +294,6 @@ class GeminiImageGenProvider(ImageGenProvider):
 
             response = requests.post(
                 endpoint,
-                params={"key": api_key},
                 headers={"x-goog-api-key": api_key, "Content-Type": "application/json"},
                 json=payload,
                 timeout=timeout,
@@ -333,11 +345,14 @@ class GeminiImageGenProvider(ImageGenProvider):
 
         image_data = None
         image_mime = "image/png"
+        response_text: List[str] = []
         candidates = body.get("candidates") if isinstance(body, dict) else None
         for candidate in candidates if isinstance(candidates, list) else []:
             content = candidate.get("content") if isinstance(candidate, dict) else None
             response_parts = content.get("parts") if isinstance(content, dict) else None
             for part in response_parts if isinstance(response_parts, list) else []:
+                if isinstance(part, dict) and isinstance(part.get("text"), str):
+                    response_text.append(part["text"].strip())
                 inline = part.get("inlineData") if isinstance(part, dict) else None
                 if isinstance(inline, dict) and isinstance(inline.get("data"), str):
                     image_data = inline["data"]
@@ -347,8 +362,12 @@ class GeminiImageGenProvider(ImageGenProvider):
                 break
 
         if not image_data:
+            detail = " ".join(text for text in response_text if text)
+            message = "Google AI Studio returned no inline image data"
+            if detail:
+                message = f"{message}: {detail}"
             return error_response(
-                error="Google AI Studio returned no inline image data",
+                error=message,
                 error_type="empty_response",
                 provider=self.name,
                 model=model,
--- a/plugins/image_gen/gemini/__init__.py
+++ b/plugins/image_gen/gemini/__init__.py
@@ -265,7 +265,11 @@ class GeminiImageGenProvider(ImageGenProvider):
             },
         }
         timeout = kwargs.get("timeout", 120)
+        timeout = kwargs.get("timeout", 180)
         try:
+            timeout = max(1.0, float(timeout))
+        except (TypeError, ValueError):
+            timeout = 180
             timeout = max(1.0, float(timeout))
         except (TypeError, ValueError):
             timeout = 120
--- a/plugins/image_gen/gemini/__init__.py
+++ b/plugins/image_gen/gemini/__init__.py
@@ -33,7 +33,7 @@ logger = logging.getLogger(__name__)
 
 BASE_URL = "https://generativelanguage.googleapis.com/v1beta"
 DEFAULT_MODEL = "gemini-3-pro-image-preview"
-_MAX_REFERENCE_IMAGES = 3
+_MAX_REFERENCE_IMAGES = 14
 _ASPECT_RATIOS = {
     "landscape": "16:9",
     "square": "1:1",
--- a/plugins/image_gen/gemini/__init__.py
+++ b/plugins/image_gen/gemini/__init__.py
@@ -260,7 +260,8 @@ class GeminiImageGenProvider(ImageGenProvider):
         payload = {
             "contents": [{"role": "user", "parts": parts}],
             "generationConfig": {
-                "responseModalities": ["IMAGE"],
+                "responseModalities": ["TEXT", "IMAGE"],
+                "imageConfig": {"aspectRatio": _ASPECT_RATIOS[aspect]},
                 "imageConfig": {"aspectRatio": _ASPECT_RATIOS[aspect]},
             },
         }
--- a/plugins/image_gen/gemini/__init__.py
+++ b/plugins/image_gen/gemini/__init__.py
@@ -32,7 +32,7 @@ from agent.secret_scope import get_secret
 logger = logging.getLogger(__name__)
 
 BASE_URL = "https://generativelanguage.googleapis.com/v1beta"
-DEFAULT_MODEL = "gemini-3-pro-image-preview"
+DEFAULT_MODEL = "gemini-3.1-flash-image"
 _MAX_REFERENCE_IMAGES = 14
 _ASPECT_RATIOS = {
     "landscape": "16:9",
--- a/plugins/image_gen/gemini/__init__.py
+++ b/plugins/image_gen/gemini/__init__.py
@@ -1,0 +1,388 @@
+"""Native Google AI Studio image generation backend.
+
+This backend calls Gemini's ``generateContent`` REST API directly with a
+``GOOGLE_API_KEY`` or ``GEMINI_API_KEY``. It intentionally does not depend on
+FAL, OpenRouter, or the Gemini text provider's OpenAI-compatible transport.
+Generated inline image data is materialized under Hermes' image cache so the
+rest of the image-generation pipeline sees the same stable path as other
+providers.
+"""
+
+from __future__ import annotations
+
+import base64
+import logging
+import mimetypes
+import os
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Tuple
+from urllib.parse import quote
+
+from agent.image_gen_provider import (
+    DEFAULT_ASPECT_RATIO,
+    ImageGenProvider,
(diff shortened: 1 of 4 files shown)


Enough1122:

**Overall:** Clean new image-generation backend plugin for direct Google AI Studio calls — well-structured, cache-integrated, and sensibly error-handled.

**What it does**
- New `plugins/image_gen/gemini/` plugin: `GeminiImageGenProvider` calls `generativelanguage.googleapis.com/v1beta` `generateContent` directly with `GOOGLE_API_KEY`/`GEMINI_API_KEY`, no FAL/OpenRouter/gateway dependency.
- Model resolution precedence: call override → `GEMINI_IMAGE_MODEL` env → scoped `image_gen.gemini.model` → top-level `image_gen.model` → default.
- Reference images (URL, data URI, or safe local path via `raise_if_read_blocked`) are base64-inlined; generated inline data is materialized through `save_b64_image` into the shared image cache so the rest of the pipeline sees the same stable path.
- Full provider interface (`list_models`, `capabilities`, `get_setup_schema`), plugin.yaml, docs table, and a solid test file (payload shape, inline reference, auth-required path, model precedence).

**Non-blocking notes**
- `modality` is set to `"image" if sources else "text"` — consistent with `image_gen_provider`'s documented contract ("which mode was used", i.e. text-to-image vs image-editing), so this is correct; just confirming it matches how the rest of the pipeline interprets the field.
- API key is passed as a URL query param (`params={"key": api_key}`), which is Google's documented REST convention but does put the key in the request line — worth being aware of for any proxy/logging in the middle; acceptable for this endpoint.
- `_load_image_gen_config()` runs on every `_resolve_model()` call; if config loading is uncached this is per-generation overhead, but config loads are generally memoized elsewhere — worth confirming it isn't a hot-path cost.
- The error taxonomy (`auth_error` on 401/403, `timeout`, `invalid_response`, `empty_response`, `io_error`) is thorough and the `_error_message` extraction handles both `error.message` and top-level `message` shapes.

No blocking issues. The provider is self-contained, follows the plugin ABC, and integrates with the existing cache and setup UX.

moonlight-lupin (commented):
## Follow-up review — field-tested sibling implementation

Thanks for this, @wesleysimplicio. We independently built a local Gemini image backend for the same need (#97474) and have been running it live since 2026-08-30. Sharing field notes plus applyable suggestions. API-behavior claims below come from live calls against `v1beta :generateContent` or current ai.google.dev docs — not from reading this diff.

### Answers to the AI-review notes above

1. **`modality` field** — correct. It follows the ABC contract ("which mode was used"). One improvement: request `responseModalities: ["TEXT", "IMAGE"]` so blocked / preamble-only responses surface their text (suggestion on the generationConfig line).
2. **Key in URL** — worth fixing. `x-goog-api-key` header, live-verified on our implementation (suggestion on the `requests.post` call).
3. **Config load cost** — non-issue. `load_config()` is cached on the config file's `(mtime_ns, size)` fingerprint and returns a deepcopy — see `_LOAD_CONFIG_CACHE` in `hermes_cli/config.py`. Per-call `_load_image_gen_config()` is a dict lookup, not disk I/O.
4. **Error taxonomy** — matches the other direct backends. 👍

### Suggested changes (inline suggestions above)

- **Auth header** — stop passing the key as a query param.
- **Catalog + default** — GA catalog is four models (`gemini-3.1-flash-image`, `gemini-3.1-flash-lite-image`, `gemini-3-pro-image`, `gemini-2.5-flash-image`); the `-preview` id is stale and `gemini-3.1-flash-image` is the sensible default.
- **TEXT + IMAGE modalities** — enables surfacing the model's text on empty/blocked responses.
- **Timeout 180 s** — pro-size renders exceed 120 s.
- **Reference cap 14** — Gemini 3 image models accept up to 14 refs (docs row should match `capabilities()`).
- **Tests** — header + modality assertions, plus invalid-JSON and empty-response coverage.

### Also worth doing on a rebase

- **Rebase onto current main.** Commit a619db66 (2026-08-28) landed a capability-gated dynamic schema plus a fleet-wide declaration⇄implementation contract test that sweeps every `plugins/image_gen/*` directory. This plugin passes it fail-closed today (no `supports_upscale` declared, none implemented), but CI on the rebased base is the proof — include `tests/tools/test_image_generate_schema.py` in the validation run.
- **Parse `usageMetadata`** (especially `candidatesTokenDetails` with `modality: IMAGE`) and surface token counts in the response `extra`. Free cost transparency; our implementation builds USD estimates from published prices on top of it — that would fit a follow-up PR.
- **Optional:** `imageConfig.imageSize` support for `gemini-3-pro-image` (1K/2K/4K), defaulting to 1K for predictable cost — mirrors the tiering pattern used in the OpenAI backend.

Happy to share our `_MODELS` / pricing tables or concrete patches for any of these.

plugins/image_gen/gemini/__init__.py
```diff
@@ -0,0 +1,388 @@
...
+            timeout = 120
+
+        endpoint = f"{BASE_URL}/models/{quote(model, safe='')}:generateContent"
+        try:
+            import requests
+
+            response = requests.post(
+                endpoint,
+                params={"key": api_key},
+                headers={"Content-Type": "application/json"},
```
moonlight-lupin: **Auth: send the key as a header, not a query param**

`params={"key": api_key}` puts the API key in the request line. Proxies, gateways, and access logs commonly capture full request lines. Google's Generative Language API accepts the `x-goog-api-key` header — we run an independent local implementation of this same backend (for #97474) with exactly this header, verified live 2026-08-30.

```suggestion
                headers={"x-goog-api-key": api_key, "Content-Type": "application/json"},
```

This also resolves the second note in the AI review above. Matching test assertion updates in the thread on the test file.

plugins/image_gen/gemini/__init__.py
```diff
@@ -0,0 +1,388 @@
...
+    resolve_aspect_ratio,
+    save_b64_image,
+    success_response,
+)
+from agent.secret_scope import get_secret
+
+logger = logging.getLogger(__name__)
+
+BASE_URL = "https://generativelanguage.googleapis.com/v1beta"
+DEFAULT_MODEL = "gemini-3-pro-image-preview"
```
moonlight-lupin: **Default model + catalog refresh**

The GA image catalog (verified against ai.google.dev, 2026-08-30) is now four models — `gemini-3-pro-image-preview` is the older preview id of the pro model:

- `gemini-3.1-flash-image` — Nano Banana 2, ~10s, best quality/cost balance — suggested default
- `gemini-3.1-flash-lite-image` — fastest and cheapest
- `gemini-3-pro-image` — premium fidelity, `imageConfig.imageSize` 1K/2K/4K, search grounding
- `gemini-2.5-flash-image` — high volume, ~$0.039/image flat

```suggestion
DEFAULT_MODEL = "gemini-3.1-flash-image"
```

Please re-verify the ids against the ai.google.dev docs before merging. Full `_MODELS` table plus pricing metadata available if useful.

plugins/image_gen/gemini/__init__.py
```diff
@@ -0,0 +1,388 @@
...
+                        "mimeType": mime,
+                        "data": base64.b64encode(data).decode("ascii"),
+                    }
+                }
+            )
+
+        payload = {
+            "contents": [{"role": "user", "parts": parts}],
+            "generationConfig": {
+                "responseModalities": ["IMAGE"],
```
moonlight-lupin: **Advertise TEXT + IMAGE modalities**

With `["IMAGE"]`-only, a safety block or preamble-only response returns a bare "no inline image data" error. With both modalities requested, the model's text part still arrives and the empty-response error can surface it (for example: `(model said: ...)` or `(blocked by safety filters)`) — much better debugging for agents.

```suggestion
                "responseModalities": ["TEXT", "IMAGE"],
                "imageConfig": {"aspectRatio": _ASPECT_RATIOS[aspect]},
```

When extracting the response, also collect `part.get("text")` from parts that have no `inlineData`, and append it to the empty-response error message. This also makes AI-review note 1 above (the `modality` field contract) unambiguous.

plugins/image_gen/gemini/__init__.py
```diff
@@ -0,0 +1,388 @@
...
+                "imageConfig": {"aspectRatio": _ASPECT_RATIOS[aspect]},
+            },
+        }
+        timeout = kwargs.get("timeout", 120)
+        try:
```
moonlight-lupin, line 286: **Raise the default timeout**

`gemini-3-pro-image` renders (1K and above, plus 4K tier) can exceed 120 s. 180 s matches the headroom other direct backends use for large renders.

```suggestion
        timeout = kwargs.get("timeout", 180)
        try:
            timeout = max(1.0, float(timeout))
        except (TypeError, ValueError):
            timeout = 180
```

plugins/image_gen/gemini/__init__.py
```diff
@@ -0,0 +1,388 @@
...
+    save_b64_image,
+    success_response,
+)
+from agent.secret_scope import get_secret
+
+logger = logging.getLogger(__name__)
+
+BASE_URL = "https://generativelanguage.googleapis.com/v1beta"
+DEFAULT_MODEL = "gemini-3-pro-image-preview"
+_MAX_REFERENCE_IMAGES = 3
```
moonlight-lupin: **Reference-image cap: 14 on Gemini 3 image models**

Gemini 3 image models accept up to 14 input images (`gemini-2.5-flash-image` is lower). Suggest a model-aware `capabilities()`; a first step is raising the constant:

```suggestion
_MAX_REFERENCE_IMAGES = 14
```

If a single constant stays, the docs table row should read "up to 14" to match `capabilities()`.

tests/plugins/image_gen/test_gemini_provider.py
```diff
@@ -0,0 +1,153 @@
...
+    assert result["model"] == "gemini-2.5-flash-image"
+    assert result["aspect_ratio"] == "portrait"
+    assert result["modality"] == "text"
+    assert Path(result["image"]).read_bytes() == b"image-bytes"
+
+    post.assert_called_once()
+    endpoint, = post.call_args.args
+    assert endpoint.endswith("/models/gemini-2.5-flash-image:generateContent")
+    assert post.call_args.kwargs["params"] == {"key": "google-test-key"}
+    payload = post.call_args.kwargs["json"]
```
moonlight-lupin, line 124: **Update request assertions for header auth + TEXT/IMAGE**

```suggestion
    assert post.call_args.kwargs["headers"]["x-goog-api-key"] == "google-test-key"
    payload = post.call_args.kwargs["json"]
    assert payload["contents"] == [{"role": "user", "parts": [{"text": "a red kite"}]}]
    assert payload["generationConfig"] == {
        "responseModalities": ["TEXT", "IMAGE"],
        "imageConfig": {"aspectRatio": "9:16"},
    }
```

Two cheap additions while here: an `invalid_response` test (non-JSON body) and an empty-response test (200 with no inline data → clear `empty_response` error).

plugins/image_gen/gemini/__init__.py
wesleysimplicio: Implemented in commit `46eedcf`. The request now asks for both `TEXT` and `IMAGE`; text parts are retained and included in the empty-image diagnostic when Gemini returns no inline image.

plugins/image_gen/gemini/__init__.py
wesleysimplicio: Updated in commit `46eedcf` and the documentation follow-up. The catalog now uses the stable IDs `gemini-3.1-flash-lite-image`, `gemini-3.1-flash-image`, `gemini-3-pro-image`, and `gemini-2.5-flash-image`, with `gemini-3.1-flash-image` as the default.

plugins/image_gen/gemini/__init__.py
wesleysimplicio: Implemented in commit `46eedcf`. The default timeout is now 180 seconds and still accepts a caller override with numeric validation and a one-second lower bound.

plugins/image_gen/gemini/__init__.py
wesleysimplicio: Addressed in commit `46eedcf`. The API key is now sent only through the `x-goog-api-key` header; it is no longer included in query parameters, which avoids exposing it in request URLs and intermediary logs.

tests/plugins/image_gen/test_gemini_provider.py
wesleysimplicio: Updated in commit `815eb99`. The tests now assert header-only authentication and TEXT+IMAGE payloads, and add invalid-JSON, empty-response, and legacy reference-limit coverage. Local validation: `12 passed` across the provider and registry suites.

plugins/image_gen/gemini/__init__.py
wesleysimplicio: Implemented in commit `46eedcf`. The provider keeps a 14-image capability for current Gemini image models and applies the documented three-image cap specifically to `gemini-2.5-flash-image`.

wesleysimplicio:
Follow-up: all six inline suggestions are addressed. I also ran `tests/tools/test_image_generate_schema.py` successfully (9 passed, 29 subtests passed). Usage metadata pricing and Pro image-size controls remain intentionally out of scope for this focused provider PR and can be follow-ups.

DanielGuru:
Applied this locally on top of current `main` to get direct Gemini image generation with our own `GEMINI_API_KEY` rather than routing through the fal / OpenRouter resellers — works well, and the provider tests are 8/8 green here. Three findings from using it in anger, all non-blocking:

**1. The docs hunk no longer applies to current `main`.** GitHub still reports this MERGEABLE, but the diff fails today:

```
$ git apply --check pr-97576.diff
error: patch failed: website/docs/user-guide/features/image-generation.md:150
error: website/docs/user-guide/features/image-generation.md: patch does not apply
```

The three code files apply perfectly cleanly with `--exclude='website/*'` — only the docs table drifted, so it is just a rebase of that one hunk.

**2. `gemini-2.5-flash-image` in `_MODELS` shuts down in 10 days.** Google's deprecations page lists its shutdown as 2026-10-02, so offering it as a selectable model will start failing shortly after merge. Minor trap if you update the list: the replacement that table *names*, `gemini-3.1-flash-image-preview`, was itself retired 2026-06-25 — the live target is `gemini-3.1-flash-image`.

**3. On the deferred `imageSize` work — the cost argument is stronger than it looks.** I measured billed output tokens directly against the API, 1:1 aspect, n=1 per cell:

| model | 1K | 2K | pixels |
|---|---|---|---|
| `gemini-3-pro-image` | 1167 | 1173 (+0.5%) | 1024² → 2048² |
| `gemini-3.1-flash-image` | 1394 | 1942 (+39%) | 1024² → 2048² |

On Pro, 2K bills essentially the same as 1K for 4× the pixels — so the current implicit-1K behaviour gives up 4× resolution for free. Flash *does* scale with size, so a per-call `imageSize` rather than a blanket default looks like the right shape.

Happy to send that as a follow-up PR on top of this one if useful — didn't want to touch yours while it's under review.

gabbyong:
Thanks @wesleysimplicio for the work on adding gemini backend, appreciate it! 
I'm working on a followup PR to make it Plugin architecture plugins/image_gen/gemini (similar plugins for oai, xai launched last week) - Dont want to touch yours too while its under review and wanted to give a heads up.

teknium1:
Thanks @wesleysimplicio — this was the first direct Gemini image backend for Hermes, and @gabbyong's #120851 (which builds on it and credits you) carried the design forward onto the `StaticImageGenProvider` plugin contract.

Closing both under one policy decision from Teknium: vendor image_gen providers (Google/Gemini included) are not added to the core tree. The right home is a standalone plugin repo plus a `plugin-catalog/<name>.yaml` entry here — the provider contract works identically from `~/.hermes/plugins/<name>` and via catalog install, and catalog PRs review fast. Full pointers (docs, submission rules, three small fixes to make first) are in the closing comment on #120851: https://github.com/NousResearch/hermes-agent/pull/120851

If you two want to co-publish that plugin, the catalog entry is the place where the credit lands. Appreciate the work here.

PR #97577, opened 2026-08-29
fix(desktop): recover group replies after renderer restart

next-devin:
## Summary

- persist a recovery marker before dispatching a group-member prompt
- reconcile interrupted replies after renderer restart with bounded polling and exact-ID deduplication
- preserve lifecycle safety across overlapping turns, room rename/disband, hydration, and plugin disposal
- keep recovery/session metadata out of gateway-visible room metadata
- surface a localized explicit unknown outcome when recovery cannot prove a reply or pass

## Validation

- `npm exec --workspace apps/desktop vitest run -- src/plugins/hermes-bots/group-turns.test.ts src/plugins/hermes-bots/group-rounds.test.ts src/plugins/hermes-bots/group-chat-view.test.ts src/plugins/hermes-bots/group-chat.test.ts src/plugins/hermes-bots/plugin-panes.test.tsx` — 150 passed
- `npm run typecheck --workspace apps/desktop`
- scoped ESLint on all changed files
- `npm run test:ui --workspace apps/desktop` — 6,732 passed across 680 files

### DIFF
--- a/apps/desktop/src/plugins/hermes-bots/group-chat-view.test.ts
+++ b/apps/desktop/src/plugins/hermes-bots/group-chat-view.test.ts
@@ -143,8 +143,17 @@ describe('disband', () => {
     room.chat.$groupChats.set({
       Gone: { log: [{ at: 2, from: { kind: 'user', name: 'You' }, id: 'g1', text: 'hello goners' }], watermarks: {} },
       Keep: {
+        holds: { remote: { at: 7 } },
         log: [{ at: 1, from: { kind: 'user', name: 'You' }, id: 'k1', text: 'hello keepers' }],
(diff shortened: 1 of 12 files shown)


dokterdok:
@next-devin, I found this draft while preparing the full Group Chat file/handoff integration in #98307. Your recovery work is the better narrow review cut, so #98307 explicitly credits and links it; the exact-ID, rename/disband, unknown-outcome, and durable owner/hold patterns are also credited in commit history. The integration draft adds file-artifact acknowledgement, attachment rollback, Stop-during-staging, quota, and queued-follow-up cases needed for end-to-end UAT; it is not intended to replace this PR. If #97577 lands first, I will drop/rebase the overlapping recovery core rather than compete with it.

Issue #56026, opened 2026-07-01
[Bug]: Weixin adapter re-delivers old messages after container restart (docker compose up without down)


895252509:
## Description

When restarting a Hermes Docker container using `docker compose up -d` (without `down` first), the Weixin adapter can re-deliver old messages from previous sessions, causing the agent to re-execute outdated instructions (e.g., answering old questions, creating duplicate kanban tasks, overwriting session titles).

## Steps to Reproduce

1. Run Hermes in Docker with Weixin (iLink bot) connected
2. Have a normal conversation session
3. Modify docker-compose.yml (e.g., add a volume mount)
4. Run `docker compose up -d` **without** `docker compose down` first
5. Observe: the agent processes old messages as new instructions

## Root Cause Analysis

The issue involves a race condition between container lifecycle and the Weixin adapter's session management:

1. **Old container** is still running, holding the Weixin bot token + platform lock
2. `docker compose up -d` starts the **new container**
3. New gateway's Weixin adapter tries to connect → fails with:
   ```
   ERROR [Weixin] Weixin bot token already in use (PID 161). Stop the other gateway first.
   ```
4. The adapter sets `_set_fatal_error(retryable=True)` and enters retry loop
5. Docker stops the old container (SIGTERM → SIGKILL after 10s)
6. Old gateway releases token + platform lock via `_release_platform_lock()`
7. New gateway retries → acquires lock → calls `getupdates` with the **persisted sync_buf**
8. **Problem:** The iLink server may have invalidated the session when the old gateway disconnected. The stale `sync_buf` cursor is no longer honored, so the API returns all pending messages including ones from previous sessions.

## Key Evidence

- Gateway logs show repeated `token already in use` errors during the overlap period
- The `sync_buf` file (`*.sync.json`) exists and has a valid value, but the server ignores it
- The `MessageDeduplicator` is purely in-memory (300s TTL) — it's empty after restart, so it can't filter out the re-delivered old messages

## Relevant Code

- `gateway/platforms/weixin.py:1282` — platform lock acquisition
- `gateway/platforms/weixin.py:1337-1384` — poll loop with sync_buf
- `gateway/platforms/weixin.py:1403-1422` — message dedup (in-memory only)
- `gateway/platforms/helpers.py:27-76` — `MessageDeduplicator` class (no persistence)
- `gateway/platforms/base.py:2578-2596` — `_acquire_platform_lock` with retryable error
[...]

Issue #36750, opened 2026-06-01
[Bug]: Weixin content dedup can drop legitimate repeated messages after the previous turn has completed


lhstack:
### Bug Description

The Weixin/WeChat gateway adapter uses content-based deduplication and in-flight content deduplication for ordinary text messages.

The expected behavior is:

- If the same message is sent again while the previous identical message is still being processed, it may be filtered or merged to avoid duplicate execution.
- If the previous identical message has already completed and the user sends the same text again, it should be treated as a new request and processed normally.

In practice, repeated ordinary text such as "怎么样了" may be silently skipped even after the previous turn has already completed or should have completed. From the user's perspective, Hermes appears to ignore the message.

This is not primarily a dangerous-command approval bug. The core problem is duplicate-message filtering: legitimate repeated user requests can be incorrectly treated as duplicates and dropped.

### Steps to Reproduce

1. Open a Weixin/WeChat DM with Hermes.

2. Send an ordinary text message, for example:
   `怎么样了`

3. While Hermes is processing, send the same text again:
   `怎么样了`

4. After the first response has completed, send the same text again:
   `怎么样了`

5. Observe whether the later repeated message is processed as a new request.

### Expected Behavior

Deduplication should distinguish between two cases:

1. **The previous identical message is still in flight**
   - The duplicate may be filtered or merged.
   - This is acceptable because it prevents duplicate execution.

2. **The previous identical message has completed**
   - The same text should be treated as a new user request.
   - It should enter `handle_message` / agent processing again.
   - It should not be silently dropped by content deduplication.

If Hermes intentionally drops a message as a duplicate, it should be visible in logs and ideally provide a user-facing acknowledgement.

### Actual Behavior

On the Weixin/WeChat platform, repeated ordinary text messages can be silently skipped.

The adapter has multiple dedup layers:

- message-id deduplication
- content fingerprint deduplication
- in-flight content deduplication

When the repeated text hits one of these dedup paths, `_process_message` returns early and the message never reaches `handle_message(event)`. No user-facing acknowledgement is sent.

This makes the user experience look like Hermes simply stopped responding.

### Affected Component

Gateway (Telegram/Discord/Slack/WhatsApp)

### Messaging Platform (if gateway-related)

WhatsApp

### Debug Report

```shell
No public debug report yet. The issue was diagnosed from local logs and source inspection.

Relevant files:

- `gateway/platforms/weixin.py`
- `gateway/platforms/base.py`
[...]

Issue #29779, opened 2026-05-21
WeChat adapter content dedup silently drops repeated commands, breaking /approve workflow


HH1162:
# WeChat adapter content dedup silently drops repeated commands, breaking /approve workflow

## Environment

- Hermes Agent: `v2026.5.16-594-g0ba7339f7`
- Platform: WeChat (iLink Bot adapter)
- Files: `gateway/platforms/weixin.py`, `gateway/platforms/helpers.py`

## Problem

The WeChat adapter has a **content fingerprint dedup** mechanism that silently discards messages with identical text from the same sender within a 5-minute TTL window, even when the user intentionally sends the same command multiple times.

### Code location

**`gateway/platforms/weixin.py` → `_process_message()` (line 1379):**

```python
# Layer 1: message_id dedup (reasonable - prevents network retry duplicates)
message_id = str(message.get("message_id") or "").strip()
if message_id and self._dedup.is_duplicate(message_id):
    return

# Layer 2: content fingerprint dedup (problematic)
text = _extract_text(item_list)
if text:
    content_key = f"content:{sender_id}:{hashlib.md5(text.encode()).hexdigest()}"
    if self._dedup.is_duplicate(content_key):
        logger.debug("[%s] Content-dedup: skipping duplicate message from %s", self.name, sender_id)
        return  # ← silently dropped, only DEBUG log
```

**`gateway/platforms/helpers.py` → `MessageDeduplicator` (line 27):**

```python
class MessageDeduplicator:
    """TTL-based message deduplication cache.

    Replaces the identical `_seen_messages` / `_is_duplicate()` pattern
    previously duplicated in discord, slack, dingtalk, wecom, weixin,
    mattermost, and feishu adapters.
    """

    def __init__(self, max_size: int = 2000, ttl_seconds: float = 300):
        # ...
```

TTL is set to `MESSAGE_DEDUP_TTL_SECONDS = 300` (5 minutes) in `weixin.py` line 96.

### Impact 1: Repeated commands are silently ignored

When a user sends the same command multiple times (e.g., `/restart-sglang` to restart a service that failed), only the first one is processed. All subsequent identical messages within the 5-minute TTL are silently discarded by the content dedup cache.

[...]

PR #97581, opened 2026-08-29
fix(weixin): dedup re-delivered messages that arrive without a messag…

xmhuangzhijun-hue:
## What does this PR do?

iLink does not always populate `message_id`, and both dedup paths are conditional on something this case lacks:

```python
message_id = str(message.get("message_id") or "").strip()
if message_id and self._dedup.is_duplicate(message_id):   # skipped when absent
    return

item_list = message.get("item_list") or []
text = _extract_text(item_list)
if text:                                                   # skipped for media
    content_key = f"content:{sender_id}:{hashlib.md5(text.encode()).hexdigest()}"
```

A message with **no `message_id` and no text** — an image, a voice note, a sticker — passes both and is processed again on every re-delivery.

Adds `_synthetic_message_id()`, used only when `message_id` is absent. It hashes transport-stable envelope fields plus the payload.

Two deliberate exclusions: `context_token` rotates by design, and unknown extension fields may rotate too. Keying on either would hand every retry a fresh identity and defeat the dedup it is meant to provide.

An ordinal field (`create_time` / `sequence` / `seq`) must be present and participates in the hash, so two legitimately identical messages from the same sender stay distinct — this does **not** widen the over-aggressive content dedup reported in #29779 and #36750. When no ordinal field is present the helper returns `""` and the caller stays on its existing path rather than risking a wrong identity.

The synthesised value also flows into `build_source(message_id=...)`, so the session and tool idempotency layers get a stable key for these messages instead of `None`.

Found on a deployment where image messages were being ingested twice.

## Related Issue

Related to #56026 (Weixin adapter re-delivers old messages after a container restart). That is a sync-cursor problem and this does not fix it — but the two compound, because the messages it re-delivers are exactly the ones this dedup currently misses. This is defence in depth, not a replacement.

Deliberately **not** touching #29779 / #36750: those ask for the text content-fingerprint dedup to be *less* aggressive. That path is untouched here.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `gateway/platforms/weixin.py` — new module-level `_synthetic_message_id()` plus `_SYNTHETIC_ID_FIELDS` / `_SYNTHETIC_ID_ORDINALS`; `_process_message` calls it only when `message_id` is empty. **Pure addition — no existing branch is modified.**
- `tests/gateway/test_weixin.py` — `TestWeixinSyntheticMessageId`, 7 cases: stable identity across re-delivery, rotating `context_token` ignored, unknown extension fields ignored, ordinal fields keep legitimate repeats distinct, payload changes change the id, and the two "no usable ordinal" cases returning `""`.

## Testing

The tests have teeth: reverting the source change fails all 7.

`pytest tests/gateway/test_weixin.py tests/gateway/test_weixin_typing.py tests/gateway/test_weixin_secret_scope.py`:

- without this change: **7 failed, 35 passed**
- with this change: **7 failed, 42 passed** (the +7 are the new tests)

The 7 pre-existing failures are unrelated (typing-ticket and secret-scope cases) and reproduce on a clean checkout of `main`.

A note for review: the field list is the part most worth a second opinion. I picked the envelope fields that looked transport-stable, but a maintainer who knows the iLink contract may want to add or remove entries — the helper is deliberately small so that list is easy to change.

### DIFF
--- a/gateway/platforms/weixin.py
+++ b/gateway/platforms/weixin.py
@@ -956,6 +956,48 @@ def _coerce_bool(value: Any, default: bool = True) -> bool:
     return default
 
 
+# Envelope fields that iLink keeps stable across a re-delivery. ``context_token``
+# is deliberately absent: it rotates by design. Unknown extension fields are
+# excluded for the same reason — a rotating field would give the same message a
+# fresh identity on every retry, which is the bug this is meant to close.
+_SYNTHETIC_ID_FIELDS = (
+    "from_user_id", "create_time", "sequence", "seq",
+    "client_id", "message_type",
+)
+# At least one of these must be present. Sender plus payload alone cannot tell a
+# resend from someone genuinely sending the same sticker twice.
+_SYNTHETIC_ID_ORDINALS = ("create_time", "sequence", "seq")
+
+
+def _synthetic_message_id(message: Dict[str, Any]) -> str:
+    """Derive a replay identity for an inbound message that carries no ``message_id``.
+
+    iLink does not always populate ``message_id``. When it is absent the primary
+    dedup is skipped, and the content-fingerprint fallback only covers text — so
+    a re-delivered image or voice note is processed again on every arrival.
+
+    Including an ordinal field keeps two legitimately identical messages from the
+    same sender distinct, so this does not widen the over-aggressive dedup
+    behaviour reported in #29779 and #36750.
+
+    Returns an empty string when no ordinal field is present, leaving the caller
+    on its existing path rather than risking a wrong identity.
+    """
+    if not any(str(message.get(key) or "").strip() for key in _SYNTHETIC_ID_ORDINALS):
+        return ""
+    identity: Dict[str, Any] = {
+        key: message.get(key)
+        for key in _SYNTHETIC_ID_FIELDS
+        if message.get(key) is not None
+    }
+    identity["item_list"] = message.get("item_list") or []
+    encoded = json.dumps(
+        identity, ensure_ascii=False, sort_keys=True,
+        separators=(",", ":"), default=str,
+    ).encode()
+    return "weixin-synthetic-" + hashlib.sha256(encoded).hexdigest()
+
+
 def _extract_text(item_list: List[Dict[str, Any]]) -> str:
     for item in item_list:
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Thoughtful fix for iLink re-deliveries that lack `message_id`: derives a stable synthetic identity from envelope fields so non-text re-deliveries (image/voice) are deduplicated, while deliberately not widening the over-aggressive dedup reported in #29779/#36750.

**What it does**
- `gateway/platforms/weixin.py` adds `_synthetic_message_id()`: if none of the ordinal fields (`create_time`, `sequence`, `seq`) is present it returns `""` (caller keeps its existing path); otherwise it hashes the stable envelope fields plus `item_list` (SHA-256 over sorted JSON) into a `weixin-synthetic-` id.
- `_process_message` falls back to this synthetic id when `message_id` is empty, so dedup runs on re-deliveries that previously slipped through.
- `context_token` is deliberately excluded (it rotates between retries) and unknown extension fields are excluded for the same reason — the tests pin both.
- Tests cover same-envelope → same id, rotating token/unknown fields → unchanged id, distinct ordinals → distinct ids, and no-ordinal → empty.

**Non-blocking notes**
- The identity includes `item_list`, so a re-delivery that changes the payload (e.g. a new `cdn_url` for the same image) yields a fresh id and is reprocessed — inherent to a content-bearing key and acknowledged by the design; just confirming that's acceptable for the intended replay window.
- `_synthetic_message_id` is a pure helper and is called on every message without `message_id`; hashing cost is negligible. `hashlib`/`json` are already imported, so no new import churn.

No blocking issues; the design (require an ordinal, exclude rotating fields) is exactly the right shape and the comment/issue links are clear.

xmhuangzhijun-hue:
Yes, treating a changed payload (including a changed `cdn_url`) as a new identity is intentional here. Without a provider-issued stable event ID, collapsing content-changing envelopes would risk suppressing a legitimate update. The helper therefore only deduplicates byte-stable re-deliveries and fails open when it cannot form a safe identity. If the iLink contract exposes a stable media/event field, that would be the better future key and the small field list is meant to make that adjustment easy.

Issue #84733, opened 2026-08-12
[Bug] Prompt-cache TTL regresses to 5m in MoA/aux, Qwen 1h marker unclamped, failover skips re-preflight


JoaoMarcos44:
## Summary

Prompt-cache TTL and stable-prefix handling degrade in auxiliary/MoA paths, and a Qwen-specific clamp is missing:

1. The main loop propagates `agent._cache_ttl` and the stable prefix (`agent/conversation_loop.py:~L2110-2124`), but the MoA/aux helper only accepts `cache_disabled` (`agent/agent_runtime_helpers.py:~L2009-2065`); MoA passes only that field (`agent/moa_loop.py:~L1738-1754`). Result: `1h` regresses to `5m` and MoA marks the whole system prompt as a breakpoint.
2. `_build_marker` always accepts `ttl:'1h'` (`agent/prompt_caching.py:~L115-120`) with no capability clamp — Qwen/Alibaba documents an explicit context cache of **five minutes** (renewed on hit), so `1h` is ignored/rejected and the expectation of a hit is false.
3. After failover to a smaller context window the compressor gets the new limit (`agent/chat_completion_helpers.py:~L2351-2378`), but the preflight is not repeated before the retry (`agent/conversation_loop.py:~L2478-2507`) — the first fallback call is still knowingly oversized.

## Impact

- More cache writes/misses and cost on MoA/aux flows.
- False 1h-cache expectations on Qwen (field ignored/rejected).
- One oversized request per provider/model failover, before any compression happens.

## Proposed fix

- Capability matrix per transport/provider/route/model with an effective TTL; propagate TTL + static prefix through the MoA/aux helper.
- Clamp Qwen to 5m (or drop the marker) until a 1h capability is proven.
- Re-run preflight after any context-window/model/provider change, before the first network call.

## Cross-references

- #79012/#79013/#79017 (prompt_cache_key scope propagation class)
- #14971 (closed: configurable prompt cache TTL)
- Docs: Anthropic 5m/1h; OpenRouter writes 5m@1.25x / 1h@2x, reads 0.1x; Qwen Alibaba context cache 5m.

kshitijk4poor:
Fixed in PR #85142 (salvage of #84782 by @webtecnica, merged). Prompt-cache TTL now propagates through MoA/aux paths, Qwen 1h markers are clamped to 5m per Alibaba's documented context-cache window, and failover re-runs the pre-API preflight against the fallback's context window.

[...]

PR #97582, opened 2026-08-29
fix(cache): preserve the configured 1h TTL on the OpenCode Go route

powriej:
## What

Restores the configured `1h` prompt-cache TTL on the **opencode-go** route. #84733 clamped
`1h -> 5m` for the whole alibaba/opencode family, reasoning from Alibaba's *published* Qwen docs.
Wire measurement on the opencode-go route contradicts the docs.

## Evidence (measured, not inferred)

Controlled run: identical request, only the `ttl` flag varying, read back after 11 minutes with
**no intervening call** — a read renews the window and would mask expiry.

| model | ttl | cache_read | result |
|---|---|---|---|
| qwen3.8-max | `1h` | 2122 | SURVIVED |
| qwen3.8-max | — | 0 | EXPIRED (control) |
| glm-5.2 | `1h` | 2092 | SURVIVED |
| minimax-m2.5 | `1h` | 0 | EXPIRED |

**The trap this route sets:** opencode-go labels *every* cache write
`cache_creation.ephemeral_5m_input_tokens` regardless of the TTL requested. That label is not
evidence of retention — it is what made the docs-based reasoning look confirmed. Retention can
only be verified by a delayed read past 5 minutes with no intervening call.

`minimax-m2.5` genuinely expires, so the allow-list is deliberately per-model, not per-provider.

## Design point worth reviewing carefully

The 1h membership is a **new, minimal allow-list** (`MEASURED_1H_PROVIDERS`) and is deliberately
**not** `ALIBABA_FAMILY_PROVIDERS`. That set also drives the cache-marker-layout opt-in in the
Anthropic cache policy — narrowing it would **disable** caching for qwen on opencode-go rather
than extend its TTL. `test_ttl_allowlist_is_separate_from_cache_layout_optin` pins the two
memberships apart so that regression cannot recur silently.

`test_every_marker_emitting_call_site_goes_through_the_central_clamp` asserts no call site
emits a marker outside the central clamp.

## Scope

Two files. No behaviour change for any provider outside the measured allow-list — `alibaba`
still clamps to `5m`, native Anthropic is unaffected.

## Tests

48/48 pass in `tests/agent/test_prompt_caching.py`.

### DIFF
--- a/agent/prompt_caching.py
+++ b/agent/prompt_caching.py
@@ -132,6 +132,64 @@ ALIBABA_FAMILY_PROVIDERS = frozenset({
 })
 
 
+# --- 1h-tier membership: an ALLOW-list, deliberately minimal ----------------
+#
+# #84733 clamped 1h -> 5m for the whole alibaba/opencode family, reasoning from
+# Alibaba's PUBLISHED Qwen docs. Wire measurement on the opencode-go route
+# contradicts the docs. Controlled run: identical request, only the ttl flag
+# varying, read back after 11 minutes with no intervening call (a read renews
+# the window and would mask expiry):
+#
+#   qwen3.8-max   ttl=1h -> cache_read 2122  SURVIVED
+#   qwen3.8-max   ttl=-  -> cache_read    0  EXPIRED    <- control
+#   glm-5.2       ttl=1h -> cache_read 2092  SURVIVED
+#   minimax-m2.5  ttl=1h -> cache_read    0  EXPIRED
+#
+# Read the two non-qwen rows for what they are: evidence about the ROUTE, not
+# about traffic Hermes sends today. anthropic_prompt_cache_policy currently
+# opts opencode-go in only for qwen models, so glm-5.2 and minimax-m2.5 on
+# that route receive no cache_control marker at all and never reach this
+# clamp in production. They constrain the route-level rule; they are not
+# live paths.
+#
+# Only opencode-go is listed: it is the only route measured. Other opencode
+# routes stay clamped because they were NOT measured, not because they are
+# known bad. opencode-zen returns cache_creation.ephemeral_1h_input_tokens for
+# Claude models, so it is a candidate -- but qwen on zen is unmeasured, so
+# adding the provider wholesale would outrun the evidence.
+#
+# WARNING: opencode-go labels EVERY write `ephemeral_5m_input_tokens` whatever
+# ttl was requested. That label is NOT evidence of the retention window -- it
+# is what made the original docs-based reasoning look confirmed. Verify only
+# with a delayed read past 5 minutes and no intervening call.
+#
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Well-evidenced, carefully-scoped repair restoring the configured `1h` cache TTL on the opencode-go route, based on delayed-read wire measurements rather than provider docs. The allow-list/deny-list design and the precedence ordering are thoughtful and extensively pinned by tests.

**What it does**
- `agent/prompt_caching.py` adds `MEASURED_1H_PROVIDERS = {"opencode-go"}` (the only measured route) and `NO_1H_TIER_MODELS = {"minimax-m2.5"}` (measured to ignore the tier even on a 1h-capable route).
- `effective_cache_ttl` now evaluates the measured-route allowance *before* the generic `is_qwen_model` clamp, returning `"5m"` only for the measured denial model, else `"1h"` on that route — while keeping every other opencode route and off-route Qwen clamped to `5m`.
- Kept separate from `ALIBABA_FAMILY_PROVIDERS` so the TTL clamp cannot silently disable cache *eligibility* (that set also drives the marker-layout opt-in); tests explicitly guard this trap.
- Extensive test coverage: precedence, negative controls, provider-spelling variants, lower-tier untouched, call-site enumeration, and honest evidence-scope notes.

**Non-blocking notes**
- `test_every_marker_emitting_call_site_goes_through_the_central_clamp` reads `agent/*.py` and `run_agent.py` source text and regex-matches call sites. This is a source-shape test (fragile in both directions: a correct refactor that renames/reformats breaks it; a wrong wiring that happens to keep the text pattern passes) — the repo guidance discourages reading source in tests. The behavioral assertion that matters is that the emitted marker actually carries `ttl: "1h"` for the deployed model, which `test_deployed_qwen37_plus_marker_carries_ttl_1h` already covers; consider dropping the source-scan or narrowing it to an explicit integration test.
- `_flat_model` lowercases and takes the segment after the last `/` — correct for aggregator-prefixed spellings; worth confirming no model id contains a `/` as a legitimate part of its name (none do today).

The repair is honest about measurement scope and keeps behavior off the measured route untouched — exactly the right shape. No blocking issues.

kshitijk4poor:
Merged via salvage PR #97708 (rebase merge — your commit landed on main as d51d66e869 with authorship preserved).

One adjustment during salvage: `test_every_marker_emitting_call_site_goes_through_the_central_clamp` read production source files with a regex and pinned an exact call-site count, which the repo's AGENTS.md bans as a change-detector/reads-source antipattern. It was replaced with a behavior-level test that drives the real `plan_cache_sections_for_destination` fan-in and asserts the emitted wire markers (ttl=1h on opencode-go, ttl stripped on unmeasured opencode) — mutation-checked. The wire-measurement evidence and the allow/deny-list design shipped verbatim.

https://github.com/NousResearch/hermes-agent/pull/97708

Thanks for the careful measurement work!

PR #97589, opened 2026-08-29
fix(web): scope duplicate memory provider fields

### LINKED ISSUE #96691: Hindsight plugin dashboard loses api_url/api_key after refresh
## Bug

The Hindsight plugin dashboard correctly persists fields that are unique to a single mode, such as `bank_id` and `bank_mission`, but `api_url` and `api_key` are lost or revert after a page refresh.

## Suspected cause

The native Hindsight plugin schema declares `api_url` and `api_key` twice: once for `cloud` mode and again for `local_external` mode, using the same field keys in both definitions.

The dashboard form appears to persist values by field key rather than by a mode-scoped identifier. As a result, the hidden duplicate fields from the inactive mode can overwrite the values entered in the visible mode during save/load.

## Reproduction

1. Open the Hindsight plugin configuration in the dashboard.
2. Select a mode that exposes `api_url` and `api_key` (for example `local_external`).
3. Enter valid values for `api_url`, `api_key`, `bank_id`, and `bank_mission`.
4. Save the configuration.
5. Refresh/reopen the dashboard.

### Actual behavior

- `bank_id` and `bank_mission` remain persisted.
- `api_url` and `api_key` are lost, cleared, or revert.

### Expected behavior

All visible configuration values for the selected mode should persist across refreshes.

## Suggested fix

Either:

- make the schema field keys unique per mode and map them back to the runtime Hindsight configuration, or
- update the dashboard form/state handling so duplicate field keys are scoped by mode and hidden fields cannot overwrite values from the active mode.

The second option would also make the dashboard more robust for other plugins that define mode-dependent fields with identical runtime keys.

icyberdeveloper:
## What does this PR do?

Fixes the web dashboard's handling of memory-provider schemas that reuse a runtime key across mutually exclusive modes. Hindsight declares `api_url` and `api_key` for both `cloud` and `local_external`; the page previously initialized the shared key from the last schema row and used only the first row when deciding whether to include it in save/setup payloads. In `local_external` mode that first row is hidden, so both values were silently dropped.

The page now keeps form state per schema row, resolves visibility against the active mode, and maps only visible rows back to runtime keys. Mode switches therefore preserve separate drafts without leaking inactive values into save/setup payloads. The same selection logic is used for normal saves and provider setup.

## Related Issue

Fixes #96691

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `web/src/pages/PluginsPage.tsx`: scope form state by schema row and submit only the active variant of each runtime key.
- `web/src/pages/PluginsPage.test.tsx`: add component regressions for mode switching, active-mode initialization, save/setup payloads, and an already-set write-only secret.

## How to Test

1. Configure the Hindsight provider in `local_external` mode.
2. Enter `api_url`, `api_key`, `bank_id`, and `bank_mission`, save, then refresh the dashboard.
3. Confirm the active mode's API URL is shown while the already-set `api_key` remains blank and carries the `set` badge.
4. Save again without entering a new key and confirm the provider remains configured.
5. Run `cd web && npm test -- src/pages/PluginsPage.test.tsx`.
6. Run `cd web && npm run check && npm run build`.

### Code

- [ ] I've run `pytest tests/ -q` and all tests pass — N/A: the repository change classifier marks this two-file web diff as `python=false`
- [x] Complete web validation passes: typecheck, 283-test Vitest suite, targeted ESLint, and production build
- [x] I've tested on my platform: Debian 13 (Linux x86_64), Node 22.23.1

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A: no user-facing workflow or config key changed
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — N/A: no config key changed
- [x] I've updated `CONTRIBUTING.md` or `AGENTS.md` if I changed architecture or workflows — N/A: no architecture or workflow changed
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — browser-only logic has no OS-specific behavior
- [x] I've updated tool descriptions/schemas if I changed tool behavior — N/A: no tool behavior changed

## Screenshots / Logs

No screenshots. The targeted component suite passes 5/5 tests, including an already-set write-only secret. A separate automated jsdom real-path E2E passed against an isolated `HERMES_HOME`, the real dashboard server, the real React page, and the real memory-provider endpoint. A literal manual graphical-browser run was not performed because this host has no usable graphical browser session.

### DIFF
--- a/web/src/pages/PluginsPage.test.tsx
+++ b/web/src/pages/PluginsPage.test.tsx
@@ -1,0 +1,316 @@
+// @vitest-environment jsdom
+import { act } from "react";
+import { createRoot, type Root } from "react-dom/client";
+import { MemoryRouter } from "react-router";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { I18nProvider } from "@/i18n";
+import type {
+  MemoryProviderConfig,
+  MemoryProviderField,
+  PluginsHubResponse,
+} from "@/lib/api";
+import PluginsPage from "./PluginsPage";
+
+const apiMocks = vi.hoisted(() => ({
+  getMemoryProviderConfig: vi.fn(),
+  getPluginsHub: vi.fn(),
+  setupMemoryProvider: vi.fn(),
+  updateMemoryProviderConfig: vi.fn(),
+}));
+const uiMocks = vi.hoisted(() => ({
+  setAfterTitle: vi.fn(),
+  showToast: vi.fn(),
+}));
+
+vi.mock("@/lib/api", () => ({ api: apiMocks }));
+vi.mock("@/plugins", () => ({ PluginSlot: () => null }));
+vi.mock("@/contexts/usePageHeader", () => ({
+  usePageHeader: () => ({ setAfterTitle: uiMocks.setAfterTitle }),
+}));
+vi.mock("@nous-research/ui/hooks/use-toast", () => ({
+  useToast: () => ({ showToast: uiMocks.showToast, toast: null }),
+}));
+
+const hub: PluginsHubResponse = {
+  orphan_dashboard_plugins: [],
+  plugins: [],
+  providers: {
+    context_engine: "compressor",
+    context_options: [],
+    memory_provider: "hindsight",
+    memory_options: [
+      {
+        available: true,
+        configured: true,
+        description: "Hindsight memory",
+        name: "hindsight",
+        status: "ready",
+      },
+    ],
+  },
+};
+
+function field(
+  key: string,
+  kind: MemoryProviderField["kind"],
+  value: MemoryProviderField["value"],
+  when: MemoryProviderField["when"] = null,
+): MemoryProviderField {
+  return {
+    description: key,
+    is_set: false,
+    key,
+    kind,
+    label: key,
+    options:
+      key === "mode"
+        ? [
+            { label: "cloud", value: "cloud" },
+            { label: "local_external", value: "local_external" },
+          ]
+        : [],
+    placeholder: "",
+    required: false,
+    url: "",
+    value,
+    when,
+  };
+}
+
+const config: MemoryProviderConfig = {
+  fields: [
+    field("mode", "select", "local_external"),
+    field("api_url", "text", "https://api.hindsight.vectorize.io", { mode: "cloud" }),
+    field("api_key", "secret", "", { mode: "cloud" }),
+    field("api_url", "text", "http://localhost:8888", { mode: "local_external" }),
+    field("api_key", "secret", "", { mode: "local_external" }),
+    field("bank_id", "text", "hermes"),
+    field("bank_mission", "text", ""),
+  ],
+  label: "Hindsight",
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Good web fix — memory-provider form state is now keyed by field *position* instead of field *key*, so provider configs that legitimately repeat a key across modes (e.g. `api_url`/`api_key` under both `cloud` and `local_external`) no longer clobber each other's drafts.

**What it does**
- `web/src/pages/PluginsPage.tsx` changes `memoryValues` from `Record<string, MemoryFormValue>` to a position-indexed `MemoryFormValue[]`, with `setMemoryFieldValue(index, value)` updating by index.
- Visibility computation is rebuilt on top of `getMemoryValuesByKey` (position → key view) and `getVisibleMemoryValues` (visible subset for save/setup), and `secretVisible` is also index-keyed.
- React keys use `${field.key}-${fieldIndex}` so duplicate-key fields render distinctly.
- New `PluginsPage.test.tsx` covers the core scenario end-to-end: editing cloud fields, switching to local_external, and verifying each mode's duplicate-key fields keep their own values through save and setup.

**Non-blocking notes**
- In `getMemoryValuesByKey`, the second `for` loop over `fields.entries()` reassigns `valuesByKey[field.key]` to exactly the same value the first `Object.fromEntries` already set — it appears redundant. If it was intended to model visibility cascades, `fieldIsVisible` already sees every field's value from the first pass, so the loop is a no-op and can be dropped (or, if a cascade is the real intent, it needs an actual re-evaluation loop to do something).
- Index-keyed `secretVisible` resets naturally when config reloads (fields re-fetch), so stale indices aren't a concern; just confirming no path reads `secretVisible` after the field list changes without a reload.
- `getInitialMemoryValues` / `getMemoryValuesByKey` accept `fields` directly — if the same field array shape is reused elsewhere, the helpers are page-local, so no coupling risk.

No blocking issues; the refactor fixes a real duplicate-key bug and the tests encode the intended scoping behavior well.

icyberdeveloper:
Thanks for the review. The second loop in `getMemoryValuesByKey` is intentional: `Object.fromEntries` collapses duplicate runtime keys to the last schema row, then the visibility pass replaces that fallback with the row active for the current mode. Without the second pass, `cloud` would inherit the later `local_external` row. The mode-switch regression covers that case.

I also added an explicit `is_set=true` regression confirming an existing secret stays blank in the UI and is submitted as blank so the backend preserves it.

icyberdeveloper:
Hi @teknium1 — could you take a look at this dashboard fix and approve the pending fork workflows (CI, Nix flake check, and Docker)? It fixes duplicate memory-provider fields across mutually exclusive modes in `PluginsPage.tsx`; the regression coverage and validation details are in the PR description. If someone else is better placed to review this area, a pointer would be appreciated. Thanks!

PR #97590, opened 2026-08-29
fix(update): retain lock across Windows shim handoff

helix4u:
## What does this PR do?

Keeps the shared update marker owned across the Windows console-shim handoff. The shim parent now transfers its claim to the spawned venv-Python child before returning the shell prompt, and the child adopts that claim until the update tail actually finishes.

Without that ownership transfer, the parent removed the marker as it exited even though the child was still updating. A second `hermes update` could then start on top of the in-flight child.

## Related Issue

Reported through Discord support; no linked GitHub issue.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- Transfer the live Windows update marker from the console shim to its venv-Python handoff child.
- Let the child adopt the transferred claim in either parent-first or child-first scheduling order.
- Leave an outer installer/orchestrator marker untouched.
- Add regressions for retained exclusion, both handoff races, outer ownership, and transfer failure fallback.

## How to Test

1. On Windows, start `hermes update` through a managed console shim and let it hand dependency sync to the venv Python child.
2. While that child is still finishing, start a second `hermes update`.
3. Confirm the second update is refused as concurrent, then confirm the marker is removed when the first update finishes.

Local verification performed: staged diff whitespace check, Python syntax parsing for all four changed files, and `scripts/check-windows-footguns.py` over the staged patch. Focused pytest was not run locally under the workstation test policy; the regression tests are included for GitHub CI.

### Code

- [ ] I've tested on my platform: Windows 11 static verification only; runtime coverage is delegated to GitHub CI

## Screenshots / Logs

N/A. This changes CLI update-lock ownership and has no visual UI change.

### DIFF
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -9340,11 +9340,23 @@ def _reexec_dependency_sync_off_windows_shim() -> bool:
     cmd = [str(python_exe), "-m", "hermes_cli.main", *sys.argv[1:]]
     if python_exe.is_file():
         try:
-            subprocess.Popen(
+            from hermes_cli.update_lock import (
+                UPDATE_LOCK_TAKEOVER_PID_ENV,
+                transfer_update_lock_to,
+            )
+
+            child = subprocess.Popen(
                 cmd,
-                env={**os.environ, _UPDATE_REEXEC_ENV: "1"},
+                env={
+                    **os.environ,
+                    _UPDATE_REEXEC_ENV: "1",
+                    UPDATE_LOCK_TAKEOVER_PID_ENV: str(os.getpid()),
+                },
                 stdin=subprocess.DEVNULL,
             )
+            if not transfer_update_lock_to(child.pid):
+                child.terminate()
+                raise OSError("could not transfer the update lock to the hand-off child")
             print(
                 f"→ Windows: {shim.name} cannot replace itself while it runs; "
                 "finishing the dependency install under the venv Python."
--- a/hermes_cli/update_lock.py
+++ b/hermes_cli/update_lock.py
@@ -46,6 +46,13 @@ Two mechanisms recognize the orchestrating parent, and either suffices:
   HANDOFF_PID_ENV export runs an old parent against a new child. Without the
   ancestry check those users get exit 2 ("Hermes is still running") on every
   GUI update forever, with no Hermes process actually running.
+
+The Windows console-shim re-exec is different: its parent must exit so the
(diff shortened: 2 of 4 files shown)


Enough1122:

**Overall:** Careful fix for the Windows console-shim re-exec: the parent now transfers its live update-lock marker to the spawned venv-Python child before exiting, closing the handoff gap where a second `hermes update` could enter while the shim parent was gone and the child hadn't claimed the lock yet.

**What it does**
- `hermes_cli/main.py:_reexec_dependency_sync_off_windows_shim` spawns the child with `UPDATE_LOCK_TAKEOVER_PID_ENV=<parent pid>`, calls `transfer_update_lock_to(child.pid)`; if transfer fails it terminates the child and raises (falls back to the non-reexec path).
- `hermes_cli/update_lock.py` adds `_takeover_pid()`, `_replace_marker_owner()` (atomic temp-file + `os.replace`, preserving unknown marker fields), and `transfer_update_lock_to()` (no-op/safe when this process doesn't own the marker — respects orchestrator/install-mode parents).
- `UpdateLock.acquire()` handles both scheduling orders: marker already names this pid (parent published it before returning) → adopt; or marker still names the takeover parent and it's a live ancestor → take over ownership. The env var is popped so it doesn't leak to grandchildren.
- Tests cover the refuse-second-update flow, orchestrator-claim preservation, child adoption of a transferred claim, winning the transfer race, and the shim re-exec paths (env passthrough, transfer call, fallback+terminate).

**Non-blocking notes**
- There is a brief window between `subprocess.Popen` and `transfer_update_lock_to`/terminate-on-failure where the child is alive but doesn't yet own the marker. On transfer failure the child is terminated, so the window is short and the failure path is loud; acceptable.
- `transfer_update_lock_to` returning True when the marker is owned by someone else (orchestrator case) means the shim parent proceeds to exit without the child owning the marker — the child's `acquire()` then relies on the ancestor/handoff check to proceed. That path is exercised only indirectly by tests; the orchestrator-preservation test covers the parent side, so behavior is pinned where it matters.

No blocking issues; the transfer is atomic, owner-aware, and both scheduling orders are handled with solid test coverage.

helix4u:
Closing as superseded by the current Windows update handoff design. Main now waits for the shim parent to exit before the child acquires its own update lock, and gateway-resume ownership follows the child. That covers the unlocked handoff tail this PR targeted through the newer implementation in 168012585e and 6f0dae8b6f.

PR #97598, opened 2026-08-29
fix(kanban): preserve worker failures and breaker trips

varijg:
## Summary

Fix two coupled Hermes Kanban failure modes discovered from repeated workers that exited without terminal task transitions:

- run dispatcher workers through the existing result-preserving `chat -Q` path so provider `rate_limit`/`billing` failures exit with sentinel 75 instead of being misclassified as clean `rc=0` protocol violations;
- make forced protocol-violation breaker trips persist a failure count at or above both the violation threshold and configured dispatcher threshold, preventing `recompute_ready()` from immediately promoting a task after `gave_up`.

## Incident classification

The affected workers never reached the model or tools. A provider HTTP 429/quota response produced zero tool calls, but the ordinary `chat -q` path returned process status 0. The dispatcher therefore classified provider failure as a clean-exit protocol violation. On retry exhaustion, `force_trip=True` persisted `consecutive_failures=1` while the effective dispatcher limit was higher, allowing immediate re-promotion.

This change does not weaken the lifecycle rule: writing `PASS.md` or returning text such as `PASS` remains nonterminal unless the worker calls `kanban_complete` or `kanban_block`.

## Validation

- focused worker/spawn/breaker/stop/single-query tests: 40 passed;
- isolated exit-75 requeue/cooldown test: 1 passed;
- total publication canary: **41 passed**;
- exact-tree independent review and publication attestation passed;
- broad Kanban subset still has 27 failures that reproduce as the identical test-ID set on the unpatched base and this branch; no full-suite green is claimed.

Additional regression coverage proves:

- ordinary and goal-mode workers each receive exactly one parser-valid `-Q`;
- the incident's actual `billing` classification maps to exit 75;
- exit 75 requeues without incrementing the failure counter and activates cooldown;
- forced trips remain blocked even with configured `failure_limit=5`;
- ordinary breaker recovery remains unchanged;
- PASS artifact/text alone remains nonterminal.

## Operational notes

- Non-quota `failed=True` workers now correctly consume the unified failure budget rather than being treated as clean-exit protocol violations.
- Quiet worker logs are intentionally thinner because interactive tool-progress callbacks are disabled.
- No schema or database migration is required; rollback is a normal commit revert plus dispatcher restart.

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -8860,7 +8860,11 @@ def _protocol_violation_streak(conn: sqlite3.Connection, task_id: str) -> int:
     return streak
 
 
-def detect_crashed_workers(conn: sqlite3.Connection) -> list[str]:
+def detect_crashed_workers(
+    conn: sqlite3.Connection,
+    *,
+    failure_limit: int = DEFAULT_SPAWN_FAILURE_LIMIT,
+) -> list[str]:
     """Reclaim ``running`` tasks whose worker PID is no longer alive.
 
     Appends a ``crashed`` event and restores the task's source phase.
@@ -9108,7 +9112,13 @@ def detect_crashed_workers(conn: sqlite3.Connection) -> list[str]:
                     conn, tid,
                     error=error_text,
                     outcome="crashed",
-                    failure_limit=violation_limit,
+                    # The violation streak decides *when* to force-trip, while
+                    # recompute_ready() uses the dispatcher's configured
+                    # failure limit to decide whether a blocked task may be
+                    # promoted. Persist at least both thresholds so a valid
+                    # forced trip stays blocked even when an operator raises
+                    # kanban.failure_limit above the violation streak bound.
+                    failure_limit=max(violation_limit, int(failure_limit)),
                     force_trip=True,
                     release_claim=False,
                     end_run=False,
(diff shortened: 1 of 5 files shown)


Enough1122:

**Overall:** Good batch of kanban resilience fixes — forced breaker trips are persisted consistently (so `recompute_ready()` can't immediately re-promote a task that was just force-blocked), and all dispatcher workers now take the fully-quiet `-Q` path so provider failures surface as real non-zero exits instead of false protocol violations.

**What it does**
- `hermes_cli/kanban_db.py:detect_crashed_workers` gains a `failure_limit` parameter (defaulting to `DEFAULT_SPAWN_FAILURE_LIMIT`) and passes `max(violation_limit, int(failure_limit))` to `_record_task_failure`, so the persisted counter reflects the stricter of the two thresholds.
- `_record_task_failure` with `force_trip=True` sets `failures = max(failures, effective_limit)` — keeping the persisted counter consistent with the terminal breaker decision so the subsequent `recompute_ready()` promotion pass can't undo the block.
- `_default_spawn` now appends `-Q` for all workers (previously goal-mode only), so rate-limit/billing exits reach the dispatcher as `EX_TEMPFAIL (75)` rather than rc=0 false protocol violations. Tests assert `-Q` on ordinary and goal-mode workers, plus quiet-mode exit-code preservation (1 / 75 for rate_limit / 75 for billing).

**Non-blocking notes**
- Making `-Q` unconditional for every worker is a behavior change beyond the fix's trigger (goal-mode was already `-Q`); it relies on the fully-quiet single-query path being correct for all workers, which the parametrized quiet-exit test covers — just confirming ordinary non-goal workers won't lose any output/telemetry they previously relied on (the per-task log redirect is preserved).
- `int(failure_limit)` is applied inside `detect_crashed_workers`; if a caller ever passes a non-numeric value this raises rather than falling back — callers pass the dispatcher's configured value, so low risk.

No blocking issues; the incident-driven reasoning (t_8dd9ecb5 forced-trip-then-immediate-promote) is well captured in the regression test.

PR #97599, opened 2026-08-29
fix(config): stop unrelated config writes from restoring stale model routing

### LINKED ISSUE #97579: [Bug]: `hermes config set` silently persists the session's model selection over the `model:` block on ANY config write (bricks profile if provider unresolvable)
## Environment

- Hermes Agent v0.20.5 (2026.8.19), git install, local `530028c2` (+1 carried commit), upstream `e60983a6`
- Linux VPS, two profiles (`<profile-a>`, `<profile-b>`), gateway + WebUI + api_server in normal use
- Provider: LiteLLM gateway on `http://<gateway-host>:4000/v1`, `model.provider: litellm`

## Summary

In v0.20.5, **any** `hermes config set <key> <value>` executed inside an agent session silently persists that session's model selection over the `model:` block in `config.yaml` — even when the requested key is completely unrelated to models. The command's output shows no warning; the model block just changes.

If the calling session was born under a broken/garbage model-selection context, this corrupts `model.provider` to an unresolvable value and **every subsequent agent init on the profile fails**.

## Repro (2 commands, deterministic)

1. In an agent session whose selected model differs from `config.yaml` `model.default` (e.g. session selected `openrouter/<gateway-model-flash>`, config default is `openrouter/<gateway-model-paid>`), run an unrelated config write:

```bash
hermes config set mcp_servers.gbrain '{"url":"http://127.0.0.1:8737/mcp","enabled":true}'
```

Output: `✓ Set mcp_servers.gbrain = ...` — no mention of models.

2. Inspect the model block:

```bash
sed -n '1,4p' ~/.hermes/<profile>/config.yaml
```

**Actual:** `model.default` changed from `openrouter/<gateway-model-paid>` to `openrouter/<gateway-model-flash>` (the session's selection).
**Expected:** only `mcp_servers.gbrain` changes; the model block is untouched.

Worse variant observed in production: a session created under a ghost provider context wrote the entire block as

```yaml
model:
  provider: custom:<ghost-provider-name>   # provider does not exist → unresolvable
  default: openrouter/<gateway-model-paid>
  key_env: ''
```

after which every agent init raised:

```
RuntimeError: Provider 'custom:<ghost-provider-name>' is set in config.yaml but no API key was found.
Set the CUSTOM_GHOST_PROVIDER_NAME_API_KEY environment variable, or switch to a different provider with `hermes model`.
```

and the gateway logged `Unknown provider 'custom:<ghost-provider-name>'` on every session.

## Production incident (2026-08-29, both profiles)

All times local (UTC+3).

- **01:55** — profile `<profile-b>` `config.yaml` model block overwritten to the ghost provider.
- **02:19:04** — an agent session in profile `<profile-a>` (spawned over the peer-DM api_server path) ran a completely benign `hermes config set mcp_servers.gbrain '{"url":"http://127.0.0.1:8737/mcp","headers":{"Authorization":"Bearer …"},"timeout":180,"connect_timeout":60,"enabled":true}'` to register an MCP server. Config mtime `02:19:06`. The same write flipped `model.provider` to the ghost provider name.
- **02:19:12+** — gateway restart picked up the corrupted config; from then on every agent init crashed with the `RuntimeError` above; WebUI streams failed with the same error; sessions created during the broken window pinned the ghost provider into their own per-session `model_config` and kept failing even after the global config was repaired.
- Repair required `hermes config set model.provider litellm`, deleting the sessions born under the broken config, and re-pinning agent cron jobs that had snapshotted the broken provider (`hermes cron edit <job> --provider litellm --model <m>`).
- Side-effect repro confirmed post-repair by re-running the same benign MCP `config set` from a session with a different model selection: `model.default` silently changed (restored immediately after).

## Impact

- A routine, unrelated config change (adding an MCP server, toggling a feature flag) can silently rewrite model routing.
- If the session's selection references a provider that no longer exists in `providers:` / `custom_providers:`, the profile is bricked for all new agent sessions until manual repair — no key, no fallback, no warning at write time.
- Cron jobs and long-lived sessions snapshot the corrupted values and keep failing after the fix, multiplying cleanup work.

## Suggested fix

1. `hermes config set` should persist **only** the requested key. Model-block stamping belongs to `hermes model` / the WebUI model picker, not to arbitrary config writes.
2. If the session-selection side effect is intentional (picker parity), gate it behind an explicit flag and print what was written (e.g. `⚠ also persisted session model selection: model.default=...`).
3. Refuse to persist an unresolvable provider: if the selected provider name can't be resolved against `providers:`/`custom_providers:` at write time, keep the existing value and warn.

## Related

- #89084 — feature request to prefer `config.yaml` model over session-persisted model on resume (same underlying mechanism, resume side)
- #91607 — `config set` splitting dotted model IDs (adjacent `model_overrides` write path)
- #94014 — api_server session-persisted model re-resolving provider

## Workaround (until patched)

Run config changes from a one-shot session pinned to the intended default (`hermes chat --model <intended-default> -q '…'`) so the side effect is a no-op, or re-verify the `model:` block after every `hermes config set` in an agent session.

fangliquanflq:
## What does this PR do?

Stops unrelated configuration writes from restoring a stale session's model routing over newer values in `config.yaml`. A long-lived session can retain an older loaded snapshot; before this change, a later full save wrote that snapshot verbatim and could undo an operator's repaired provider while preserving no warning that model routing had changed.

### Symptom

After `hermes config set mcp_servers.gbrain ...`, `model.provider` or `model.default` can revert to the values held by an older session snapshot. If that snapshot names an unavailable custom provider, new agent and gateway initialization fails until the profile configuration is repaired.

### Impact

Operators can lose a repaired model route while changing an unrelated setting. The stale provider can then prevent new sessions from resolving credentials or initializing a provider.

### Bug Cause

**Trigger:** `hermes_cli/config.py:save_config` saves a mutable snapshot returned by `load_config()` after another writer has changed `config.yaml`.

**Causal chain:**
1. A long-lived session loads configuration containing its then-current model route.
2. Another writer repairs the model route and `hermes config set` adds an unrelated key.
3. The long-lived session later saves its older full snapshot, replacing the repaired model route and the newly added key with stale values.

**Why it is wrong:** `save_config()` could not distinguish caller changes from untouched values inherited from an old load, so every value in the stale snapshot won over newer on-disk state.

**Working sibling / contrast:** Direct `set_config_value()` writes only the requested key and does not read session model state. The corruption occurs when a later full-snapshot save overwrites that correct single-key write.

**Ruled out:** Session model environment variables do not alter the direct `hermes config set` path. Running the reported CLI and Console commands with different `HERMES_MODEL` and `HERMES_INFERENCE_MODEL` values preserved the existing model block.

### Fix

Track the baseline on mutable configurations returned by `load_config()` and perform a three-way merge during `save_config()`: caller changes win, while untouched paths take their latest on-disk values. Serialize single-key and full-document writes with the shared config lock, preserve baseline tracking through normal `.copy()` calls, and keep the tracked mapping compatible with safe YAML serialization.

## Related Issue

Closes #97579

## Type of Change

## Changes Made

- `hermes_cli/config.py` - track loaded baselines and merge stale snapshots without clobbering newer values.
- `cli.py` - serialize runtime single-key writes with the shared configuration lock and invalidate caches eagerly.
- `tests/cli/test_cli_save_config_value.py` - cover stale writes, copied snapshots, YAML compatibility, and the reported unrelated `config set` model-routing regression.

## How to Test

1. Load a configuration snapshot containing provider `openrouter`.
2. Change `model.provider` to `litellm`, add `mcp_servers.gbrain`, then save the older snapshot with an unrelated agent change.
3. Verify the saved file retains `litellm`, retains `mcp_servers.gbrain`, and includes the intended agent change.
4. Run the relevant test files:

```bash
scripts/run_tests.sh tests/cli/test_cli_save_config_value.py -q
scripts/run_tests.sh tests/hermes_cli/test_config.py -q
```

Result: 125 tests passed. A baseline run of the focused regression failed with `openrouter` restored instead of `litellm`; the fixed tip passes.

### Code

- [x] I've run the repository test entry point on the relevant tests and all tests pass

- [x] Documentation update is N/A; behavior and regression coverage are self-contained
- [x] Cross-platform impact considered; the merge and locking behavior is platform-independent
- [x] Tool description and schema updates are N/A; no model tool changed

## Screenshots / Logs

N/A. The focused regression test captures the stale provider restoration before the fix and the preserved routing after the fix.

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -4918,13 +4918,21 @@ def save_config_value(key_path: str, value: any) -> bool:
     config_path = get_hermes_home() / 'config.yaml'
     
     try:
-        # Ensure parent directory exists (for ~/.hermes/config.yaml on first use)
-        config_path.parent.mkdir(parents=True, exist_ok=True)
-        
-        # Save back atomically while preserving comments, ordering, quotes, and
-        # readable Unicode in user-edited config.yaml.
-        from utils import atomic_roundtrip_yaml_update
-        atomic_roundtrip_yaml_update(config_path, key_path, value)
+        from hermes_cli import config as config_module
+
+        with config_module._CONFIG_LOCK:
+            # Ensure parent directory exists (for ~/.hermes/config.yaml on first use)
+            config_path.parent.mkdir(parents=True, exist_ok=True)
+
+            # Save back atomically while preserving comments, ordering, quotes, and
+            # readable Unicode in user-edited config.yaml.
+            from utils import atomic_roundtrip_yaml_update
+            atomic_roundtrip_yaml_update(config_path, key_path, value)
+
+            # The atomic replace invalidates stat-keyed caches naturally, but
+            # discard the raw cache eagerly for coarse-mtime filesystems.
+            config_module._RAW_CONFIG_CACHE.pop(str(config_path), None)
+            config_module._LOAD_CONFIG_CACHE.pop(str(config_path), None)
         
         # Enforce owner-only permissions on config files (contain API keys)
         try:
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -318,6 +318,12 @@ class _LoadedConfig(dict):
     def __init__(self, value: Dict[str, Any]):
         super().__init__(value)
         self._baseline = copy.deepcopy(value)
+
+    def copy(self) -> "_LoadedConfig":
+        """Return a shallow snapshot without losing stale-write tracking."""
+        cloned = _LoadedConfig(dict(self))
+        cloned._baseline = copy.deepcopy(self._baseline)
+        return cloned
 # Env var names written to .env that aren't in OPTIONAL_ENV_VARS
 # (managed by setup/provider flows directly).
 _EXTRA_ENV_KEYS = frozenset({
@@ -387,6 +393,11 @@ _EXTRA_ENV_KEYS = frozenset({
     "COPILOT_ACP_BASE_URL",
 })
 import yaml
+
+yaml.SafeDumper.add_representer(
+    _LoadedConfig,
+    yaml.representer.SafeRepresenter.represent_dict,
+)
 
 from hermes_cli.colors import Colors, color
 from hermes_cli.default_soul import DEFAULT_SOUL_MD, is_legacy_template_soul
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -400,8 +400,10 @@ def _represent_loaded_config(dumper, value):
     return dumper.represent_dict(value)
 
 
-yaml.SafeDumper.add_representer(_LoadedConfig, _represent_loaded_config)
-yaml.Dumper.add_representer(_LoadedConfig, _represent_loaded_config)
+for _dumper_name in ("SafeDumper", "Dumper", "CSafeDumper", "CDumper"):
+    _dumper = getattr(yaml, _dumper_name, None)
+    if _dumper is not None:
+        _dumper.add_representer(_LoadedConfig, _represent_loaded_config)
 
 from hermes_cli.colors import Colors, color
 from hermes_cli.default_soul import DEFAULT_SOUL_MD, is_legacy_template_soul
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -2709,17 +2709,16 @@ def _merge_loaded_config_snapshot(base: Any, requested: Any, current: Any) -> An
             return _CONFIG_MISSING
         return copy.deepcopy(current)
 
-    if (
-        isinstance(base, dict)
-        and isinstance(requested, dict)
-        and isinstance(current, dict)
+    if isinstance(base, dict) and isinstance(requested, dict) and (
+        isinstance(current, dict) or current is _CONFIG_MISSING
     ):
+        current_mapping = {} if current is _CONFIG_MISSING else current
         merged: Dict[str, Any] = {}
-        for key in base.keys() | requested.keys() | current.keys():
+        for key in base.keys() | requested.keys() | current_mapping.keys():
             value = _merge_loaded_config_snapshot(
                 base.get(key, _CONFIG_MISSING),
                 requested.get(key, _CONFIG_MISSING),
-                current.get(key, _CONFIG_MISSING),
+                current_mapping.get(key, _CONFIG_MISSING),
             )
             if value is not _CONFIG_MISSING:
                 merged[key] = value
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -2705,6 +2705,8 @@ def _merge_loaded_config_snapshot(base: Any, requested: Any, current: Any) -> An
         return _CONFIG_MISSING
 
     if base is not _CONFIG_MISSING and requested == base:
+        if current is _CONFIG_MISSING:
+            return _CONFIG_MISSING
         return copy.deepcopy(current)
 
     if (
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -304,6 +304,20 @@ _RAW_CONFIG_CACHE: Dict[str, Tuple[int, int, Dict[str, Any]]] = {}
 # calls read_raw_config. Also covers mutation of the module-level cache
 # dicts above.
 _CONFIG_LOCK = threading.RLock()
+
+
+class _LoadedConfig(dict):
+    """Mutable config snapshot carrying the state it was loaded from.
+
+    ``save_config`` uses the baseline for a three-way merge so a caller that
+    saves an older snapshot does not overwrite keys changed on disk since its
+    ``load_config`` call. It remains a dict subclass to preserve the public
+    return contract and normal deepcopy behavior.
+    """
+
+    def __init__(self, value: Dict[str, Any]):
+        super().__init__(value)
+        self._baseline = copy.deepcopy(value)
 # Env var names written to .env that aren't in OPTIONAL_ENV_VARS
 # (managed by setup/provider flows directly).
 _EXTRA_ENV_KEYS = frozenset({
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** Fixes stale-session config writes restoring old model routing by introducing a tracked `_LoadedConfig` dict subclass whose baseline powers a three-way merge in `save_config`. Well-tested and clearly explained.

**What it does**
- `hermes_cli/config.py:load_config` now returns `_LoadedConfig` (a dict subclass) that carries `_baseline` — a deepcopy of the snapshot as loaded.
- `save_config` detects a tracked snapshot, reloads the *current* disk state, and runs `_merge_loaded_config_snapshot(base, requested, current)`: paths the caller changed win, untouched paths inherit current (including concurrent additions/deletions), and caller-level deletions are preserved.
- `cli.py:save_config_value` now holds `_CONFIG_LOCK` around the atomic write and eagerly evicts `_RAW_CONFIG_CACHE` / `_LOAD_CONFIG_CACHE` (for coarse-mtime filesystems).
- YAML representers registered for `_LoadedConfig` (Safe/Dumper/CSafe/CDumper) so it stays serializable; 6 new tests cover stale-snapshot scenarios, nested deletions, parent-section deletion, `.copy()` retention, and YAML round-trip.

**Non-blocking notes**
- `save_config_value` acquires `_CONFIG_LOCK`, but this diff does not show `save_config` (the full-write path) taking the same lock, so the lock only serializes single-key writes against each other. The three-way merge still guards correctness across a `save_config_value` + `save_config` race, but if you want strict write/write serialization, consider taking `_CONFIG_LOCK` in `save_config` too.
- List-valued leaves always resolve to `copy.deepcopy(requested)` (no per-element merge), so a stale caller holding an older list can still clobber a concurrent list edit. Likely acceptable given config lists are rare and element-identity merging is ambiguous — worth a comment in code so it's a conscious choice.
- `_LoadedConfig._baseline` only tracks the snapshot at `load_config()` time; if the same object is mutated across several `save_config` calls, the baseline is re-set to the last *requested* snapshot after each save, which is the intended "reused stale object" behavior — correct, just subtle.

fangliquanflq:
Thanks for the detailed review. I checked each note against the committed tip:

- No lock change is needed: `save_config()` already holds `hermes_cli.config._CONFIG_LOCK` for its complete read/merge/write sequence, and `save_config_value()` now acquires that same lock. The focused lock regression also passes.
- List handling is intentional. An untouched list (`requested == base`) inherits the current on-disk value; only a true same-leaf conflict uses the documented "caller changes win" rule. Element-wise list merging would require an identity/order contract that these configuration lists do not have, so I am keeping leaf-level conflict resolution.
- Reused tracked objects are already covered by the baseline update at the end of `save_config()`: it records the caller's requested snapshot rather than the merged disk result, preventing preserved concurrent values from becoming caller edits on a later save.

No code changes were needed. I reran:

- `scripts/run_tests.sh tests/cli/test_cli_save_config_value.py -q` - 13 passed
- `scripts/run_tests.sh tests/hermes_cli/test_config.py -q` - 114 passed

teknium1:
Thanks for iterating on this — the lock/list reasoning checks out.

Parking with needs-repro before deciding on the approach. Re-reading #97579: the on-disk shape after the clobber (`key_env: ''`, `provider: custom:<ghost>`) looks like a full `model` block written by a process that held a stale snapshot (TUI `/model` persist path in `tui_gateway/server.py`, a plugin calling `save_config`, gateway restart), not something `hermes config set` produces — `set_config_value` is already key-scoped. If we can identify that writer, the smallest fix is making that one call site key-scoped (~10 LOC), rather than making every `load_config()` return a change-tracked dict.

@fangliquanflq / reporter: which Hermes processes were running when `model.provider` flipped (gateway? TUI/Desktop? a plugin?), and did any of them do a `/model` switch or settings save around that time? With that we can pick between the targeted fix and this merge-aware `save_config`.

fangliquanflq:
Thanks — the issue report narrows the runtime context but does not identify the full-document writer conclusively.

- It explicitly says the gateway, WebUI, and `api_server` were in normal use, and the 02:19 write was initiated from an agent session over the peer-DM `api_server` path.
- It does not say that the TUI/Desktop was running, or that a `/model` switch, WebUI settings save, gateway model switch, or plugin save occurred around the flip.
- At the reported `e60983a7` baseline, `tui_gateway/server.py::_persist_model_switch` already used targeted `save_config_value()` calls. The two gateway `/model` persistence paths in `gateway/slash_commands.py` still used full `save_config()` calls, but they require an explicit model-switch action that the report does not establish.
- `hermes config set` itself is key-scoped, so the current evidence supports a later or concurrent stale full save, not direct model stamping by that command. The focused regression demonstrates that stale-full-save class, but it cannot identify which production caller ran.

I cannot provide the missing process/action timeline beyond the reporter's issue body, so I agree the targeted-writer choice should wait for the reporter's logs or confirmation. No code change or test rerun in response to this comment.

teknium1:
Thanks for the careful work here — closing after a full caller audit against current `main` and v0.20.5 (530028c2):

- Every `save_config` writer that touches `model.*` (`set_config_value`, `save_config_value` used by CLI/TUI `/model`, gateway `/model` persist, web_server model assignment / custom-endpoint / `PUT /api/config`, `model_switch`) does load → mutate → save inside one function or is key-scoped. No caller holds a `load_config()` snapshot across a long-lived process and saves it later, so the stale-snapshot class this PR guards against has no instance in the tree.
- The reported on-disk shape (`key_env: ''` + `provider: custom:<ghost>`) has no writer: every `model.key_env` write is truthiness-guarded. It is not producible by `hermes config set` (which never reads `model` state — the PR body notes the same).
- Making every `load_config()` return a tracked dict changes semantics for ~150 `save_config` callers ("untouched path takes disk value"), which would stop wizard flows from deliberately re-saving a loaded value.

If a real stale writer or a cross-process TOCTOU race surfaces, the right fix is key-scoped at that site or a file lock on `set_config_value`/`save_config_value` (~10 LOC). Asked the reporter on #97579 for the carried-commit diff, config before/after, and running processes; happy to reopen with a repro.

PR #91567, opened 2026-08-21
feat(buzz): continue mentioned reply threads

vrtak-ai:
## Summary
- Continue direct replies in a Buzz thread after an initial `@mention`, while keeping mentions required elsewhere.
- Add `thread_require_mention: true` for strict per-reply mention gating.
- Handle out-of-order Nostr delivery with bounded, expiring pending replies and rewind poll/WebSocket cursors.
- Document the setting and add regression coverage for continuation, ordering, expiry, and long reply chains.

## Verification
- `scripts/run_tests.sh tests/gateway/test_buzz_adapter.py tests/gateway/test_buzz_websocket.py` (33 passed)
- `.venv/bin/ruff check plugins/platforms/buzz/adapter.py tests/gateway/test_buzz_adapter.py`
- `git diff --check`
- Independent review: passed (no blocking security or logic issues).

PR #84492, opened 2026-08-12
fix: preserve Buzz reply threads

tomwattsca:
## Summary
- preserve Buzz channel reply threads by propagating the NIP-10 root event into `MessageSource.thread_id`
- prefer an explicit `root` marker and otherwise use the first positional `e` tag
- start a new thread for a top-level channel post while leaving DMs unthreaded

## Root cause
Inbound Buzz events were dispatched without thread context. The send path therefore had no `thread_id` and posted every response as a new top-level message.

## Tests
- `uv run --with pytest --with pytest-asyncio python -m pytest tests/gateway/test_buzz_adapter.py -o 'addopts=' -q`
- 32 passed

The tests cover top-level posts, single and nested replies, explicit NIP-10 markers, malformed tags, DMs, and ensuring nested replies remain on the original root rather than creating sub-threads.

PR #97608, opened 2026-08-29
fix(buzz): reply at channel level with --broadcast instead of nesting threads

robottwo:
## Summary
- Pass `--broadcast` alongside `--reply-to` on all three Buzz send paths (`send()`, `send_image()`, standalone media sender) so agent replies keep NIP-10 linkage but render as top-level timeline entries.
- Previously every adapter reply nested as a collapsed thread under the user's message.

## Root cause
The buzz CLI emits a NIP-10 `e`-tag reply for `--reply-to` with no `broadcast` tag. Both Buzz clients classify such messages as thread replies:
- Desktop: `desktop/src/features/messages/lib/threading.ts` — `isThreadReply()` returns true when a parent `e`-tag exists **and** `isBroadcastReply()` (a `["broadcast","1"]` tag) is false.
- Mobile: `mobile/lib/features/channels/timeline_message.dart` — main timeline includes a message only when `parentId == null || _isBroadcastReply(msg)`.

Native `buzz-acp` agents reply with the broadcast tag, which is why native agents' answers appear at channel level while adapter replies nested. Adding the tag aligns the adapter with native behavior; reply linkage (notifications, thread summary counts) is preserved by the `e`-tag.

## Verification
- `venv/bin/python -m pytest tests/gateway/test_buzz_adapter.py -o addopts= -q` — 23 passed
- `venv/bin/python -m pytest tests/gateway/test_buzz_websocket.py -o addopts= -q` — 4 passed
- Client-behavior claims verified against Buzz Desktop and Mobile source (linked above).

Related but distinct: #84492 (inbound thread context propagation — why replies had `--reply-to` targets at all) and #91567 (mention gating for thread continuation).

### DIFF
--- a/plugins/platforms/buzz/adapter.py
+++ b/plugins/platforms/buzz/adapter.py
@@ -613,7 +613,12 @@ class BuzzAdapter(BasePlatformAdapter):
         args = ["messages", "send", "--channel", str(chat_id), "--content", "-"]
         reply_target = reply_to or (metadata or {}).get("thread_id")
         if reply_target:
-            args += ["--reply-to", str(reply_target)]
+            # Reply with --broadcast so the message keeps its NIP-10 reply
+            # linkage (notifications, thread summary) but renders at channel
+            # level instead of nesting into a collapsed thread. Both Buzz
+            # Desktop and Mobile treat a ["broadcast","1"] tag as a
+            # top-level timeline entry.
+            args += ["--reply-to", str(reply_target), "--broadcast"]
         code, out, err = await self._run_cli(args, input_text=content)
         if code != 0:
             return SendResult(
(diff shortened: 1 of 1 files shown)


Enough1122:

**Overall:** Fixes Buzz agent replies nesting as collapsed threads by adding `--broadcast` alongside `--reply-to` on all three send paths, so messages keep NIP-10 reply linkage but render at channel level. Minimal, consistent change across `send()`, `send_image()`, and `_standalone_send()`.

**What it does**
- `send()`: appends `--broadcast` to the args when `reply_target` is set.
- `send_image()`: appends `--broadcast` to the args when `reply_to` is set.
- `_standalone_send()`: appends `--broadcast` when `thread_id` is set.

**Non-blocking notes**
- Relies on the buzz-cli `messages send` accepting `--broadcast` on all three paths; worth a quick confirm that the standalone/`--file` media path accepts the flag the same way the text path does (it's the same subcommand here, so likely yes).
- No tests changed — reasonable since this is CLI-flag wiring, but a mocked `_run_cli` assertion on the exact arg list per path would lock the behavior in.

teknium1:
Closing — unconditional --broadcast removes threading for every deployment rather than making it a choice. The configurable opt-out that landed via #99429 (platforms.buzz.reply_to_mode: off / extra.reply_in_thread: false) achieves the flat-channel outcome per deployment. Thanks for pushing on the pain point; the config route keeps both camps working.

PR #88892, opened 2026-08-18
fix(curator): stop teaching the model a delete the guard always refuses

liuhao1024:
## What does this PR do?

Aligns the curator consolidation prompt with the fail-closed delete guard. `_curator_consolidation_delete_guard` rejects a background-pass `skill_manage(action="delete")` whose `absorbed_into` is omitted **or empty** — a bare prune is refused to keep the skill active (#29912's fail-open lesson), and the deterministic inactivity prune is the only legitimate prune path. But the prompt (`agent/curator.py`, `CURATOR_REVIEW_PROMPT`) still instructed the model: *"or `absorbed_into=\"\"` when you're truly pruning with no forwarding target"* — so an obedient model made exactly the call the guard deterministically refuses (#88882).

The prompt now advertises only declared-consolidation deletes: `MUST pass absorbed_into=<umbrella>` (umbrella must already exist on disk), states that a delete without a verified consolidation target is **rejected fail-closed**, and points no-target pruning at the deterministic inactivity prune instead. The `_extract_absorbed_into_declarations` docstring is updated to describe the guard contract rather than the retired instruction — the parser itself still accepts empty-string values so pre-guard curator runs and non-background callers keep classifying correctly.

No behavior change to the guard or the tool — this is the prompt/tool contract becoming self-consistent.

## Related Issue

Fixes #88882

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `agent/curator.py`
  - `CURATOR_REVIEW_PROMPT` delete bullet: removed the empty-`absorbed_into` instruction; now states the umbrella requirement, the fail-closed rejection, and the inactivity-prune outlet for no-target pruning.
  - `_extract_absorbed_into_declarations` docstring: describes the guard contract; notes empty-string values remain parsed for backward compat.
- `tests/agent/test_curator_classification.py`
  - New `TestConsolidationPromptMatchesDeleteGuard`: the prompt never advertises an empty-target delete; it states the fail-closed rejection, the inactivity-prune outlet, and the `MUST pass absorbed_into=<umbrella>` requirement.

## How to Test

1. `python -m pytest tests/agent/test_curator_classification.py tests/tools/test_skill_manager_tool.py -q`
   Observed result: 62 passed.
2. `python -m pytest tests/agent/ -q -k curator`
   Observed result: 59 passed, 4 skipped.
[...]

Issue #88882, opened 2026-08-18
Curator prompt still instructs direct pruning after background deletes were prohibited


22373448:
The background curator intentionally prohibits deleting a skill without a verified consolidation target. _curator_consolidation_delete_guard() rejects both an omitted and empty absorbed_into:

tools/skill_manager_tool.py:474

```python
declared = isinstance(absorbed_into, str) and absorbed_into.strip()
if declared:
    return None

return {
    "success": False,
    "_fail_closed": True,
}
```

However, the active curator prompt still tells the LLM to perform direct pruning with an empty target:

agent/curator.py:547

```python
"skill_manage action=delete — archive a skill. MUST pass "
"`absorbed_into=<umbrella>` when you've merged its content into another "
"skill, or `absorbed_into=\"\"` when you're truly pruning with no "
"forwarding target."
```

As a result, when consolidation is enabled and the LLM follows the prompt, a direct-prune call deterministically fails:

```python
skill_manage(
    action="delete",
    name="stale-skill",
    absorbed_into="",
)
```

> Note: This issue was identified by an automated testing tool for academic research and manually verified. If you have any concerns about this type of reporting, please let me know, and I will adjust my workflow accordingly.

andrexibiza:
Topology correction against live `main@b1ff8722a53ee223485ac9804945acf07ef5c601`: this defect is **not** actually closed on main yet.

Current `agent/curator.py` still contains the exact contradictory instruction recorded here — `absorbed_into="" when you're truly pruning with no forwarding target` — and its extractor docstring still describes empty-target pruning as valid. The issue is mechanically `closed / completed`, but the source postcondition does not match that state.

Open #88892 (`liuhao1024`, head `06c866d8754cba431c7f351c17bc0f0904ad18d4`) remains the originating narrow implementation of this prompt/guard alignment. #97609 is now the broader consolidation carrier: exact head `9a8e18710478b0fe2cbc4f8b9075bd3a3d552014` absorbs the #88892 contract while also removing ambient terminal authority, pinning the curator to a static+dispatch-whitelisted skill surface, moving cron-reference migration into the delete settlement path, and binding rollback to the exact archive generation.

[...]

Issue #29912, opened 2026-05-21
Curator may archive active skills during umbrella pass without verified consolidation (fail-open behavior)


sanchomuzax:
## Bug description
The curator can archive currently active/operationally critical skills during umbrella consolidation, even when evidence for successful consolidation is missing or inconsistent.

In our run, 10 operational skills were archived in one pass, including:
- `facebook-high-signal-routine`
- `linkedin-high-signal-routine`
- `google-photos-memory-ops`
- `cron-run-error-recovery`

This breaks active workflows because the original skill names are removed from normal lookup and the replacement umbrella may not be a drop-in equivalent.

## Evidence
Curator report:
- `~/.hermes/logs/curator/20260521-132901/REPORT.md`
- `~/.hermes/logs/curator/20260521-132901/run.json`

Key contradictions from `run.json`:
- `counts.before: 10`, `counts.after: 0`, `archived_this_run: 10`
- `consolidated_this_run: 0`
- `cron_jobs_rewritten: 0`
- `pruned[*].source: "fallback (model named missing umbrella, no tool-call evidence)"`

So all candidates were archived, but no measured consolidations happened and no cron rewrites were applied.

## Reproduction (high-level)
1. Have a set of agent-created skills that are still referenced in real operational routines.
2. Run curator pass with consolidation logic enabled.
3. Observe full-wave archive into `~/.hermes/skills/.archive/`.
4. Observe active skill names no longer available under normal skill lookup path.

## Actual behavior
- Curator archives a whole cluster of active skills in one run.
- Report summary says umbrella consolidation happened.
- Machine counters show zero consolidations and fallback/no-evidence pruning paths.
- Cron references are not rewritten (`cron_jobs_rewritten: 0`).

## Expected behavior
- Never archive skill names that are currently in active use unless safe migration is verified first.
- If umbrella creation fails or is not evidenced, **fail closed**: do not archive originals.
- Require hard validation before archive:
  - replacement skill exists and passes lookup,
  - reference rewrites (cron + other references) succeeded,
  - post-check proves no unresolved references.
- If any check fails, keep originals active and emit warning-only report.

## Impact
High operational risk: active automations lose known skill names and can silently degrade/fail.

## Suggested fix directions
1. Add pre-archive guard: block archive when `consolidated_this_run == 0` for planned absorption set.
2. Add mandatory reference rewrite/validation gate before delete/archive.
[...]

Issue #26655, opened 2026-05-15
fix(curator): LLM consolidation pass uses skill_manage delete instead of terminal mv, bypassing .archive/ recovery


WegoW:
## Problem

The Curator's LLM consolidation pass instructs the agent to "archive" skills, but the LLM repeatedly uses `skill_manage(action='delete', absorbed_into='')` instead of `terminal` with `mv` to move the skill into `~/.hermes/skills/.archive/`.

This means "archived" skills are **permanently deleted** from the filesystem and are **not recoverable** via `hermes curator restore <name>`. They only survive in pre-run backups (tar.gz), and the `.archive/` directory is left empty even though the report claims skills were archived.

Additionally, the LLM pass has **no awareness of `archive_after_days: 90`** from the Curator config. It can archive (or delete) skills that are only days old, bypassing the configured timeout entirely.

This was observed with both **kimi-k2.6** and weaker models (e.g. gemini-2.5-flash-lite) on the Curator's consolidation pass.

## Root cause

The `CURATOR_REVIEW_PROMPT` (`agent/curator.py`, line 329+) tells the LLM:

- Rule 2: "DO NOT delete any skill. Archiving (moving to .archive/) is the maximum destructive action"
- But the LLM is given `skill_manage action=delete` as a valid tool and repeatedly uses it instead of `terminal mv`
- No mention of `archive_after_days` or `stale_after_days` in the prompt instructing the LLM to respect configured timeouts

## Expected behavior

1. LLM should only use `terminal` to `mv` skill directories into `.archive/`, never `skill_manage delete`
2. LLM should be instructed not to archive any skill younger than `archive_after_days` (configurable, default 90)
3. Optional: Config flag like `curator.llm_pass_enabled: false` to disable LLM pass while keeping auto-transitions

## Suggested fix

- Add `archive_after_days` and `stale_after_days` context to `CURATOR_REVIEW_PROMPT` with hard instruction not to touch younger skills
- Change tool instructions: only `terminal mv` for archiving, remove `skill_manage delete` from allowed tools during consolidation pass (or add a guard that intercepts delete and converts to mv)
- Consider adding `curator.llm_pass_enabled` config key

6tizer:
## Additional case: Curator consolidated skills outside its candidate list

We hit this same bug with an extra dimension: the LLM consolidated skills that were **not in the curator-managed candidate list** at all.

### What happened (2026-05-17 run)

[...]

PR #97609, opened 2026-08-29
fix(curator): prevent terminal archive bypass

trevorswanson:
## Summary

The curator consolidation fork was given both `skills` and `terminal` toolsets. Its prompt allowed direct filesystem moves into `.archive/`, which bypassed the guarded `skill_manage(delete, absorbed_into=...)` path and therefore bypassed cron skill-reference migration.

In the observed Aug. 28 run, 20 skills were archived while `cron_jobs_rewritten` remained `0`; `note-taking/obsidian` was subsequently left referenced by enabled cron jobs. The source run report is local evidence and is not included in this PR.

This change removes `terminal` from the curator fork and updates the prompt so all skill archival goes through `skill_manage`. That keeps archive and dependent-reference migration on the guarded path.

## Changes

- Restrict the curator LLM fork to the `skills` toolset.
- Remove prompt instructions that suggest shell-based skill moves.
- Require archival through `skill_manage(action=delete, absorbed_into=...)`.
- Update regression coverage for the reduced tool surface.

## Validation

```text
python3 -m pytest -q tests/agent/test_curator.py tests/cron/test_rewrite_skill_refs.py
43 passed in 2.37s
git diff --check
```

## Related upstream issues

This is a distinct bypass, related to but not duplicating:

- #29912 — curator archived active skills without verified consolidation.
- #26655 — curator archive/delete path bypassed recoverable `.archive/` semantics.
- #88882 — curator prompt remained inconsistent with the fail-closed delete guard.

## Notes

The account used for this PR has read-only permission on the upstream repository, so a separate upstream issue could not be opened. The distinct issue report and reproduction evidence are included here for maintainer triage.

Credit: #88892 (@liuhao1024)


andrexibiza (commented):
Reviewing exact head `e8cab9481780aaf4c400cffd70a0ccad2b4b7d00` against merge base `ac6c8028e00d01ee2f299ba7fd03329c7f10382d`; live `main` is `62e8126c6929d1a5d9fd650317214ef240767d4e`, so this branch is currently 1 commit ahead / 1 behind. I also checked the current `skill_manage` guard/archive path, the actual tool-resolution path, curator report/cron migration, the existing review state, exact-head hosted CI, and the relevant curator lineage (#29912, #26655, #88882/#88892, #96962/#97045/#96980/#95449, and #23398).

The direction is good: taking raw `terminal` away from an autonomous curator is materially stronger than relying on prompt discipline, and routing intended mutations through `skill_manage` restores provenance, validation, archive semantics, and the background-review mutation guards. There are three boundary holes I think need to be closed before this head lands.

### P1 — the live prompt still teaches a delete that the guard deterministically refuses

This head removes the terminal escape hatch, but leaves this exact live instruction in `CURATOR_REVIEW_PROMPT`:

```text
absorbed_into="" when you're truly pruning with no forwarding target
```

Current `tools/skill_manager_tool.py::_curator_consolidation_delete_guard()` does the opposite for this actor: while `is_background_review()` is true, an omitted **or empty** `absorbed_into` is rejected fail-closed; only a non-empty declared umbrella is admitted. That guard is intentional #29912 behavior — no verified replacement, no archive.

So after this PR removes `terminal`, a model that follows the remaining prune instruction has no valid effect path: the prompt asks it to call the only mutation API in a shape that the mutation boundary always refuses. This is not hypothetical; #88882 records the same prompt/guard contradiction, and open #88892 already carries the complementary fix in these same `agent/curator.py` lines.

Please make the actor contract singular: this LLM pass may delete only as a verified consolidation into an existing umbrella; no-target pruning belongs exclusively to the deterministic inactivity path. Add a cross-contract regression that exercises the real prompt plus the guard, not just one side. Because #88892 already contains that prompt/guard alignment, either sequence it before this branch or absorb that exact contract here while preserving its contributor credit and making the supersession explicit rather than silently recreating the work.

### P1 — dependent-reference migration is still post-effect and best-effort; `skill_manage(delete)` is not the settlement boundary

The PR description/comments say forcing archive through `skill_manage` keeps “archive and dependent-reference migration on the guarded path.” The first half is true; the second is not true in the current architecture.

`skill_manage(action="delete", absorbed_into=...)` validates the declared destination and performs the recoverable archive, but it does not rewrite cron references. `agent/curator.py` derives the `consolidated_map` later and then calls `cron.jobs.rewrite_skill_refs(...)` **after the destructive LLM pass**. That block is explicitly best-effort: the code catches an exception, logs it, records `cron_rewrites.error`, and continues. A failed rewrite can therefore leave the source skill archived while an enabled cron still points at the removed identity — the same unsafe terminal state this PR is motivated by, reached through a different edge.

That is precisely the stronger invariant recorded in #29912: replacement exists, dependent references migrate successfully, and the post-check is clean **before** the old identity is allowed to disappear; otherwise keep the original active. The terminal bypass made that failure easier, but removing terminal does not make settlement atomic.

Please add an adversarial witness where an enabled cron references the source, the consolidation has a valid `absorbed_into`, and `rewrite_skill_refs` fails (or its postcondition is not satisfied). The accepted result must not be “source archived + stale enabled reference.” Either move reference migration/validation into the consolidation mutation transaction, or compensate by rolling the archive back before reporting success. #23398 remains a complementary residual on freeform `job["prompt"]` references even after structured `skill`/`skills` migration; this PR does not need to solve that whole issue, but it should not claim the dependent-reference class is guarded until the structured settlement is actually fail-closed.

### P1 — `enabled_toolsets=["skills"]` is not an exact capability allowlist, and the new test deliberately avoids the live widening path

The safety argument here is “there is no shell/filesystem tool in this fork.” The production resolution path does not prove that invariant.

The new test computes:

```python
resolve_toolset("skills", include_registry=False)
```

but `AIAgent` ultimately calls `get_tool_definitions(enabled_toolsets=["skills"])`, and `model_tools` resolves that with the normal `resolve_toolset(toolset_name)` path. The normal path includes registry-provided/plugin toolset members; `model_tools` even documents that plugin-registered tools are resolved through that same path. A plugin/overlay can therefore register another tool into the `skills` toolset and widen the curator's callable surface beyond the three built-ins even though this static test remains green.

That matters here because the entire repair is an authority reduction. A semantic bucket whose membership is extensible is not a durable proof that this autonomous actor has exactly `skills_list`, `skill_view`, and `skill_manage`.

Please pin the curator to an exact tool-name capability set at the execution boundary (or otherwise resolve a static-only surface and enforce the same set at dispatch), then add a poisoned-registry regression: register a fake extra tool into `skills`, construct the real curator fork, and prove the extra tool is neither advertised nor invocable. The current `include_registry=False` documentary check is useful, but it proves the static catalog rather than the runtime authority surface.

### Interlocks / merge order / acceptance

#97045 is overlapping/competing work on the same raw-terminal mutation defect, not a duplicate to merge blindly alongside this one: it keeps terminal read-only while adding a sink-side skills-tree write guard and changes the same curator prompt. #97609 chooses the smaller actor-capability shape by removing terminal outright. Pick one canonical ownership boundary and preserve the other contributor's root-cause/guard work in the supersession record. #96980 is complementary recovery for already-incomplete archive/ledger entries; #95449 is complementary because it makes the guarded `skill_manage` path usable across worker-context snapshots. #26655 is historical lineage, but its old recommendation to use raw `terminal mv` predates the current recoverable `skill_manage` archive semantics and should not be revived.

Acceptance is independently not green at this exact object. Hosted runs for `e8cab9481780aaf4c400cffd70a0ccad2b4b7d00` are `action_required` for CI (`33229451545`, zero jobs), Docker (`33229451076`), and Nix (`33229450815`); the later label-rerun workflows are also `action_required`. The reported local `43 passed` is useful evidence, but it is not exact-head hosted acceptance. This head also needs to be restacked over current `main@62e8126c...` and re-proved there.

There is a strong repair here: remove ambient shell authority, force skill mutations through the typed owner, and keep the original package when safe re-homing cannot be completed. Once the prompt matches the actual guard, the dependent-reference settlement is fail-closed, the capability surface is exact rather than registry-widenable, and the restacked commit is green, this becomes the kind of boundary hardening the curator needs. 🚀

trevorswanson:
Thanks for the thorough review — all three P1s are addressed in 46662d5. Responding to each separately:

---

### 1. Prompt-Guard Contradiction

You're right, and it's worse than a wording nit: `_curator_consolidation_delete_guard()` already fail-closes on any `skill_manage(delete)` call from the background-review actor that omits `absorbed_into` or passes it empty (that guard predates this PR — it was added for #29912). The prompt just never got updated to match, so it was instructing the model down a path that was already dead on arrival.

Fixed by reworking `CURATOR_REVIEW_PROMPT` so the LLM pass is told, explicitly, that it has no path to a bare prune: `skill_manage(delete)` always requires a real `absorbed_into=<umbrella>`, full stop. Where the prompt previously said "pass `absorbed_into=""` when truly pruning," it now says: don't call `skill_manage` at all in that case — list the skill under `prunings` in the structured YAML summary as a **recommendation**, and let a human or the separate deterministic inactivity sweep act on it. The `prunings` list keeps its name and shape (so downstream parsing/reporting is unaffected) but its documented meaning changes from "archived with no merge target" to "flagged, not archived."

Net effect: the curator's LLM pass can no longer prune outright — only consolidate. That's not a new restriction introduced here; it's just now truthfully documented. (Also answering your implicit question: this doesn't reduce what the curator can actually do in production, since the guard was already blocking it.)

### 2. Incomplete Transaction Safety (post-effect reference migration)

Agreed this was the real gap. Moved cron-reference migration out of the end-of-run batch pass and into `_delete_skill` itself (`tools/skill_manager_tool.py`): for a verified consolidation, `rewrite_skill_refs(consolidated={name: absorbed_target})` now runs synchronously, in the same call, immediately after the archive succeeds — not after the whole curator run completes.

If that rewrite raises, we don't leave the skill archived with a dangling cron reference: we call `restore_skill(name)` to undo the archive and return `success: False` so the caller can see the delete didn't actually take effect. If the rollback *itself* fails (belt-and-suspenders case), the error says so explicitly and flags it for manual attention rather than pretending everything's fine.

The old end-of-run batch call in `agent/curator.py` stays in place as an idempotent backstop (it still matters for reconciliation-only findings the tool-call audit catches that the model didn't declare), but it's no longer the primary mechanism and the comment there now says so.

Added `test_verified_consolidation_migrates_cron_refs_in_same_transaction` and `test_failed_cron_migration_rolls_back_the_archive` in `tests/tools/test_skill_manager_tool.py` covering both the happy path and the rollback.

### 3. Non-Exact Capability Allowlist (registry widening)

Confirmed and fixed at the root rather than patching around it. `enabled_toolsets=["skills"]` was resolving via the normal `resolve_toolset(name, include_registry=True)` path, which merges in anything a plugin/overlay registered into the "skills" toolset — exactly the gap you called out, and exactly why the static test (`include_registry=False`) wasn't representative of the runtime call site.

Added a `restrict_toolsets_to_static` flag threaded through `AIAgent.__init__` → `init_agent` → `get_tool_definitions` → `resolve_toolset(include_registry=False)`, and set it `True` on the curator's forked agent. Production tool-schema resolution for this fork now goes through the exact same static-only path the test already asserted against, so a plugin registering a new tool into "skills" can no longer reach the curator.

On top of that, added a dispatch-time thread-local tool whitelist around the fork's `run_conversation()` call — the same mechanism `agent/background_review.py` already uses for the memory/skill review fork (#15204). This is defense-in-depth: even if some future change to tool assembly let a non-whitelisted name reach the model, the dispatch gate refuses to execute it, and the whitelist is cleared in a `finally` so it can't leak onto other tool calls on the same thread.

New tests: `test_restrict_toolsets_to_static_blocks_registry_widened_skills_toolset` (registers a fake tool into "skills" via the real registry and proves it's excluded under the new flag, whereas it's included without it — closes the exact gap the old documentary-only test couldn't catch), and `test_review_fork_installs_and_clears_runtime_tool_whitelist`.

---

### CI / restack

CI is currently showing `action_required` on this head — from what I can see that looks like a workflow-approval gate rather than a code issue on my end; might need a maintainer to approve the run. Separately, I checked whether the branch needed restacking: `fix/curator-safe-consolidation`'s merge-base with current `main` is `main`'s tip, so it's already fully rebased — nothing to do there.

### Overlap with #97045

Confirmed this is real, not a false positive — #97045 is addressing a different angle of the same underlying issue class (curator's `terminal` access letting it bypass the skill ledger, there for support-file demotion rather than whole-skill archival), and it touches the same `enabled_toolsets` line and similar prompt text. I didn't try to resolve that here since picking canonical ownership between the two PRs isn't mine to decide unilaterally — flagging it back to you/the maintainers rather than guessing.

All changes are in commit `46662d5` on this branch. Full test suites for the touched modules (`tests/agent/test_curator.py`, `tests/tools/test_skill_manager_tool.py`, `tests/test_model_tools.py`, `tests/test_toolsets.py`, `tests/run_agent/test_background_review_toolset_restriction.py`) pass.

andrexibiza (commented):
Re-reviewing exact head `46662d58488479545d9d205297e67e2373e46305` after the three prior boundary fixes.

The prompt/guard contradiction is closed, the registry-widening capability hole is closed with static-only resolution plus a dispatch-time whitelist, and cron-reference migration is now moved into the delete path with compensation on migration failure. The poisoned-registry witness is the right shape.

One rollback correctness hole remains in that new compensation path.

### P1 — rollback is name-based, not bound to the archive generation this delete just created

`archive_skill(name)` deliberately handles an already-existing `.archive/<name>` by moving the current active package to `.archive/<name>-<timestamp>`. But when `rewrite_skill_refs(...)` raises, `_delete_skill()` calls `restore_skill(name)` and treats any `True` result as proof that *this* archive was rolled back.

`restore_skill(name)` does not have that identity. It searches an exact `.archive/<name>` first and only falls back to timestamp-suffixed copies if no exact-name candidate exists. So this sequence is legal today:

1. an older `.archive/foo/` already exists;
2. current active `foo` is archived by this consolidation as `.archive/foo-<timestamp>/`;
3. cron migration fails;
4. `restore_skill("foo")` restores the *older* `.archive/foo/` into the active slot;
5. `_delete_skill()` returns `success: false` saying the archive was rolled back and `foo` is active again, while the just-archived current generation is still in `.archive/foo-<timestamp>/` and the active bytes have silently regressed to the older copy.

That violates the exact postcondition the compensation is supposed to establish. A failed consolidation may not replace the current package with a different archived generation and then report successful rollback.

Please bind rollback to the exact archive artifact created by this mutation instead of re-resolving by semantic name. The clean shape is for `archive_skill` to return an exact rollback handle/path (or expose a bounded inverse that takes that exact artifact) and for `_delete_skill` to restore only that object. Add the adversarial regression with a pre-existing exact-name archived copy and byte-distinct current package: force `rewrite_skill_refs` to fail, then prove the current package's exact bytes are restored, the older archive remains untouched, and no timestamped copy of the current generation is stranded.

Hosted acceptance is also still non-evidence on this two-commit train: exact head `46662d58…` has CI `33231060115`, Docker `33231059715`, and Nix `33231059720` all `action_required`. The predecessor `e8cab948…` was likewise never executed. Every surviving commit still needs real green execution (or the reviewed tree needs rematerialization as one coherent commit and that exact object executed).

andrexibiza (commented):
Re-reviewing exact head `9a8e18710478b0fe2cbc4f8b9075bd3a3d552014` after the rollback-binding repair.

The prior generation bug is closed on this object. `_delete_skill()` no longer re-resolves rollback by semantic skill name: `archive_skill_with_path()` returns the exact archive artifact created by this mutation, and failure compensation uses `restore_skill_from_path(name, archive_dest)`. The adversarial collision witness is also now the right one: an older `.archive/narrow/` is kept byte-distinct and untouched while the current generation is forced into the timestamped collision path, cron migration fails, and the test proves the CURRENT bytes—not the stale archive—are restored with no name-based substitution. The earlier prompt/guard contradiction, registry-widened `skills` capability surface, and post-effect cron migration holes also remain closed here.

One repository-acceptance blocker remains before this train is ready to land:

### P1 — this repair still grows `agent/curator.py` past the 2K boundary instead of moving the authority into a bounded owner

This PR modifies `agent/curator.py` in hunks that already reach past line 2,000 (`@@ -1969,... +2003,...` and later). The current Hermes architecture gate is explicit: no modified file may remain above 2,000 lines; godfiles are sharded and never restored. This change is exactly an authority-boundary repair—curator fork construction, its exact tool capability, consolidation prompt, and post-run interpretation now form a coherent unit—so keeping that logic embedded in the oversized orchestrator makes the safety boundary harder to audit and leaves future fixes colliding in the same godfile.

Please extract the bounded curator review-fork authority (at minimum the fork capability/prompt construction and its directly coupled helpers) into a sub-2K module, leave `agent/curator.py` as orchestration/delegation, and move the relevant regressions with the owner. The extraction must be behavior-preserving: exact static `skills` resolution + dispatch whitelist, consolidation-only delete contract, transactional cron migration, and exact-generation rollback all need to remain pinned through the public call path.

Landing evidence is independently still incomplete. All three surviving commits are unexecuted by hosted PR CI: `e8cab948…` has CI `33229451545`, Docker `33229451076`, Nix `33229450815`; `46662d58…` has `33231060115` / `33231059715` / `33231059720`; head `9a8e1871…` has `33251157628` / `33251157323` / `33251157298`; every one is `action_required`. I attempted the exact-head CI rerun again and GitHub returned 403 `Resource not accessible by integration`, so that execution boundary is external. The source/placement boundary above is not external and still needs to be completed before acceptance.

Current `main@b1ff8722a53ee223485ac9804945acf07ef5c601` has moved since the recorded base. The only submitted-path overlap I found in that interval is `run_agent.py`, and the main-side edit is in the session flush scan around line 2229 while this PR's edit is the curator-fork construction near the earlier initialization path; GitHub currently reports the PR mergeable/rebaseable. Still, the eventual restacked/rematerialized object must be re-executed, not inherit this head's evidence.

Topology/credit also needs to preserve #88892 / `liuhao1024`: that PR is the originating public implementation of the prompt-vs-delete-guard alignment now absorbed here. #97609 is the broader consolidation carrier, but the predecessor should be recorded as provenance rather than silently recreated.

trevorswanson:
Addressed both new items in `d8d8c46`, plus a clarification on the CI status.

### File-size gate (`agent/curator.py` past 2,000 lines)

Confirmed — the consolidation-guard fix in this branch pushed `agent/curator.py` to 2101 lines. Rather than trim comments to squeeze under the line, I extracted the actual review-fork subsystem into a new bounded module, `agent/curator_review_fork.py`: `CURATOR_REVIEW_PROMPT` / `CURATOR_DRY_RUN_BANNER`, provider/model resolution (`_resolve_review_runtime` / `_resolve_review_model`), and the forked-`AIAgent` construction and execution (`_run_llm_review`) — including the toolset restriction and dispatch-time whitelist wiring from the earlier rounds. `curator.py` keeps scheduling, state, the deterministic lifecycle transitions, and report writing/rendering.

- `agent/curator.py`: 2101 → 1603 lines
- `agent/curator_review_fork.py`: 538 lines (new)

`curator.py` re-exports the moved names (`from agent.curator_review_fork import CURATOR_DRY_RUN_BANNER, CURATOR_REVIEW_PROMPT, _resolve_review_model, _resolve_review_runtime, _run_llm_review`), so every existing `curator.X` / `from agent.curator import X` reference — including the test suite's monkeypatches of `_run_llm_review` and its `caplog`-based check on the deprecated-config warning — keeps working with no test changes needed beyond binding the new module's logger explicitly to `"agent.curator"` (not `__name__`) so log filtering by that name is unaffected by the split. Full suite still green (176 passed; the 3 known-unrelated sandbox gaps — `fastapi`/`uvicorn` and a `web_search` API key — are unchanged from before this PR).

### CI still `action_required`

Worth flagging in case this wasn't visible on your end: this is GitHub's own gate for a first-time/external contributor's workflow run — it requires a maintainer to click "Approve and run workflows" before CI actually executes, for any of the three commits so far. It's not something fixable from this side (fork CI config, code changes, or a rebase don't touch it) — it just needs a maintainer with write access on `NousResearch/hermes-agent` to approve the pending run on the latest commit. `.github/workflows/ci-review-comment.yml` in this repo says as much directly, in its own trigger comment: *"a first-time contributor's run sits in `action_required` until a maintainer approves it."* Once approved, CI will run and evaluate the actual current head rather than any earlier evidence.

### Provenance — #88892 / @liuhao1024

Confirmed and appreciate the flag. #88892 independently fixed the same prompt-vs-delete-guard contradiction this PR's first round addressed (round 1, commit `46662d5`) — same root cause, arrived at separately. Recording that here so it's not lost: **the prompt/guard-alignment fix in this PR overlaps with, and should credit, #88892 by @liuhao1024 as the originating public implementation.** I don't have write access to edit that PR or this one's description from here, so flagging it for whoever merges to add the credit line (or to close #88892 with a reference to wherever the credited fix lands) — happy to add a "Credit: #88892 (@liuhao1024)" line to this PR's description myself if that's the preferred place to record it.

Mergeable/rebaseable state is unaffected by any of this — no new conflicts introduced.

---
_Generated by [Claude Code](https://claude.ai/code)_

andrexibiza (commented):
Re-reviewed exact head `d8d8c46ebbd0c1598a7c524c3b0b72ee3b94727e` after the extraction commit.

The blocker from my previous review is fixed as far as `agent/curator.py` itself is concerned: the LLM review fork now has a bounded owner in `agent/curator_review_fork.py`; `agent/curator.py` is down to ~1.6K lines and the new owner is ~538 lines. I re-read the moved runtime path, not just the commit message. The static `skills` surface, dispatch-time whitelist, background-review origin, provider/runtime resolution, prompt contract, and cleanup still survive the split. The earlier prompt/guard, transactional cron-reference migration, registry-widening, and exact-generation rollback fixes also remain present.

One acceptance blocker remains from the **full submitted FILE-LIST**, and this corrects the scope of my prior 2K finding: this PR also modifies `run_agent.py` and `agent/agent_init.py`. On this exact head, `run_agent.py` still has live source beyond line 3500 and `agent/agent_init.py` beyond line 3000. Under the repository's 2K invariant, extracting only `curator.py` does not close the placement gate while the same change adds `restrict_toolsets_to_static` plumbing through two existing godfiles.

Please finish that ownership move rather than leaving new security-boundary plumbing inline in those files: give the static-toolset / tool-surface policy a bounded owner and keep `AIAgent` / `init_agent` as compatibility delegates only. The regression contract should remain end-to-end: a poisoned registry must still leave the curator fork advertising and dispatching exactly `skills_list`, `skill_view`, and `skill_manage`, and the dispatch whitelist must still clear after the fork exits.

I also re-ran the landing edge against current `main@e387cbc0aa0fc89560bc14438762c7722663db05`. Main is 75 commits beyond the recorded base. Three submitted path names are touched in that interval (`run_agent.py`, `tools/skill_manager_tool.py`, `tests/tools/test_skill_manager_tool.py`), but the actual PR hunk regions I checked remain intact on current main, and GitHub still reports this head mergeable/rebaseable; the PR is blocked by acceptance/check state rather than a textual merge conflict.

Execution is still not transferable: all four surviving commits remain without a green CI/Docker/Nix matrix. Current head is `action_required` at CI `33253798715`, Docker `33253798456`, and Nix `33253798442`. I attempted the exact CI rerun and GitHub rejected it with 403 `Resource not accessible by integration`.

So: the previous curator extraction blocker is closed, the functional security fixes remain intact, but the **full-FILE-LIST 2K placement gate** and hosted every-commit execution gate are still open.

trevorswanson:
Addressed both new items in `d8d8c46`, plus a clarification on the CI status.

### File-size gate (`agent/curator.py` past 2,000 lines)

Confirmed — the consolidation-guard fix in this branch pushed `agent/curator.py` to 2101 lines. Rather than trim comments to squeeze under the line, I extracted the actual review-fork subsystem into a new bounded module, `agent/curator_review_fork.py`: `CURATOR_REVIEW_PROMPT` / `CURATOR_DRY_RUN_BANNER`, provider/model resolution (`_resolve_review_runtime` / `_resolve_review_model`), and the forked-`AIAgent` construction and execution (`_run_llm_review`) — including the toolset restriction and dispatch-time whitelist wiring from the earlier rounds. `curator.py` keeps scheduling, state, the deterministic lifecycle transitions, and report writing/rendering.

- `agent/curator.py`: 2101 → 1603 lines
- `agent/curator_review_fork.py`: 538 lines (new)

`curator.py` re-exports the moved names (`from agent.curator_review_fork import CURATOR_DRY_RUN_BANNER, CURATOR_REVIEW_PROMPT, _resolve_review_model, _resolve_review_runtime, _run_llm_review`), so every existing `curator.X` / `from agent.curator import X` reference — including the test suite's monkeypatches of `_run_llm_review` and its `caplog`-based check on the deprecated-config warning — keeps working with no test changes needed beyond binding the new module's logger explicitly to `"agent.curator"` (not `__name__`) so log filtering by that name is unaffected by the split. Full suite still green (176 passed; the 3 known-unrelated sandbox gaps — `fastapi`/`uvicorn` and a `web_search` API key — are unchanged from before this PR).

### CI still `action_required`

Worth flagging in case this wasn't visible on your end: this is GitHub's own gate for a first-time/external contributor's workflow run — it requires a maintainer to click "Approve and run workflows" before CI actually executes, for any of the three commits so far. It's not something fixable from this side (fork CI config, code changes, or a rebase don't touch it) — it just needs a maintainer with write access on `NousResearch/hermes-agent` to approve the pending run on the latest commit. `.github/workflows/ci-review-comment.yml` in this repo says as much directly, in its own trigger comment: *"a first-time contributor's run sits in `action_required` until a maintainer approves it."* Once approved, CI will run and evaluate the actual current head rather than any earlier evidence.

### Provenance — #88892 / @liuhao1024

Confirmed and appreciate the flag. #88892 independently fixed the same prompt-vs-delete-guard contradiction this PR's first round addressed (round 1, commit `46662d5`) — same root cause, arrived at separately. Recording that here so it's not lost: **the prompt/guard-alignment fix in this PR overlaps with, and should credit, #88892 by @liuhao1024 as the originating public implementation.** I added a "Credit: #88892 (@liuhao1024)" line to this PR's description.

Mergeable/rebaseable state is unaffected by any of this — no new conflicts introduced.

andrexibiza (commented):
Re-reviewed exact current head `c4a9188cf4289df8e300003f601f4c56d97c3924` after the placement repair and upstream integration.

The full-FILE-LIST blocker from my `d8d8c46…` review is closed on this object. The submitted file list is now eight paths and no longer includes `run_agent.py` or `agent/agent_init.py`; the static-toolset policy is contained in `model_tools.py`, while the curator fork itself remains in bounded `agent/curator_review_fork.py`. The runtime path still pins the actor to the static `skills` definitions, overwrites both advertised tools and `valid_tool_names`, installs the dispatch-time thread whitelist, and clears it after the fork exits. The poisoned-registry regression exercises the real widened registry path and proves the extra tool does not survive the static restriction.

I also rechecked the mutation settlement path. Consolidation archives return the exact destination artifact, structured cron references are rewritten synchronously inside the delete operation, and migration failure compensates with `restore_skill_from_path(name, archive_dest)` rather than a name search. The collision regression keeps an older `.archive/narrow/` byte-distinct and proves rollback restores the current generation while leaving the stale archive untouched. The consolidation-only prompt contract remains aligned with the fail-closed delete guard; no-target pruning is recommendation-only in this LLM pass.

GitHub currently reports the PR open, non-draft, and mergeable. I do not see a remaining source/placement blocker in the submitted tree.

Hosted acceptance is still not settled. This is now a seven-commit submitted train, and exact current head `c4a9188…` has CI `33256647401`, Docker `33256647046`, and Nix `33256647042` all `action_required`. A train whose head has not executed cannot satisfy the every-surviving-commit gate. The remaining landing boundary is therefore hosted execution/approval, not another source defect; no green evidence should be transferred from any predecessor or local run.

trevorswanson:
Thanks for all the reviews. Having cleared the blockers and issues you've identified, I believe we need to wait for a maintainer to approve running GH actions. Neither of us can run those checks in the repo because we lack permission.

It's GitHub's first-time-contributor approval gate on the upstream repo (confirmed from .github/workflows/ci-review-comment.yml's own trigger comment in this repo: "a first-time contributor's run sits in action_required until a maintainer approves it"). No branch state, rebase, or code change on this side flips that — it needs a maintainer with write access on NousResearch/hermes-agent to click "Approve and run workflows" on the current head. Once that happens once, subsequent pushes to this PR should auto-run without needing repeat approval (that's the normal GitHub behavior for a trusted contributor within a PR), so this shouldn't need re-raising after the first approval.

egilewski:
too large to review safely

This PR changes 1298 production lines before tests and docs. Please split it or add a focused justification if it should stay together.

Signed: GPT-5.6-luna-high in Codex

trevorswanson:
> too large to review safely
> 
> This PR changes 1298 production lines before tests and docs. Please split it or add a focused justification if it should stay together.
> 
> Signed: GPT-5.6-luna-high in Codex
> 

Justification: the original, functional changes made by the PR touch significantly fewer production lines of code, but due to files crossing the 2000-line godmode limit, additional forks had to be made which artificially inflates the line count.

kshitijk4poor:
Closing as **superseded** — the prevention this PR targets landed on main via #99491.

Commit `37ae0e1d39` (merged 2026-08-31) removes `terminal` from the consolidation fork entirely (`enabled_toolsets=["skills"]`), which closes the archive-bypass class by construction — no terminal channel, no bypass — and `98e2f110cf` adds the full-package rollback restore. Issue #96962 is closed.

Your work here mattered to that outcome: #99491's body credits you as the first to propose removing terminal from the fork and for documenting the cron-migration bypass evidence that drove the cluster. The guard-layer approach in this branch was reviewed seriously (five andrexibiza rounds), but the ~1300-line guard surface lost to the ~90-line channel removal on the "smallest fix that closes the whole bug class" principle.

Thanks @trevorswanson for the depth of investigation and the responsiveness across review rounds.

PR #97610, opened 2026-08-29
feat(buzz): presence, typing, seen reactions, and NIP-AO observer telemetry

robottwo:
Stacks on #97608 (the `--broadcast` reply fix).

## Summary
- **Presence (kind 20001)**: publish `online` at connect + 60s heartbeat, `offline` on graceful disconnect. Desktop renders the agent as an online member (relay presence TTL is 180s; native buzz-acp agents do the same).
- **Typing (kind 20002)**: indicator every 3s while drafting + a 5s keepalive loop for long tool-heavy turns; auto-clears client-side 8s after the last indicator. Sending the reply cancels the loop.
- **Seen acknowledgment**: 👀 reaction on the inbound message at dispatch, removed when the reply lands — read-receipt parity with native agents.
- **Observer telemetry (NIP-AO, kind 24200)**: NIP-44-encrypted `ObserverEvent` frames carrying `session/update` payloads (thinking chunks, tool status) to the owner's pubkey, which Buzz Desktop renders as the live "thinking" transcript for the agent. This is the same wire format `buzz-acp` emits for remote agents — Hermes speaks it directly, no ACP harness required.
- **`nip44.py`**: dependency-free NIP-44 v2 encrypt/decrypt (pure-Python secp256k1 ECDH x-coordinate, HKDF-SHA256, ChaCha20 stream, HMAC-SHA256, v2 payload framing), ported from the nostr-rs reference implementation, with round-trip tests.

## Design notes
- Observer owner resolution: `BUZZ_OWNER_PUBKEY` (hex) or the existing `BUZZ_AUTH_TAG` NIP-OA tag's second element. Unset/invalid → telemetry silently no-ops; presence/typing/reactions work regardless.
- All telemetry is fire-and-forget and never raises into the send path.
- Frames reuse the adapter's existing NIP-42-authenticated ephemeral WebSocket.

## Known limitation (relay-side, not fixable client-side)
Buzz relays gate kind 24200 on an `agent_owner_pubkey` DB mapping (see `buzz-relay/src/handlers/event.rs` `handle_agent_observer_event`). That mapping is only materialized when a **non-member** agent authenticates via NIP-OA whose owner **is** a member (`MembershipDecision::ViaOwner`), or on open relays. A direct member agent on a closed relay — the standard Hermes fleet setup — gets frames rejected with `restricted: observer frame is not authorized for this agent owner`. Presence/typing/reactions are unaffected. We plan to file this upstream with Block; the client-side work here is complete and correct against the wire format.

## Verification
- `venv/bin/python -m pytest tests/gateway/test_buzz_adapter.py tests/gateway/test_buzz_websocket.py -o addopts= -q` — 27 passed
- Owner-resolution unit checks (env unset / `BUZZ_OWNER_PUBKEY` valid+invalid / `BUZZ_AUTH_TAG`): all pass
- NIP-44 round-trip against the Python port; NIP-42 auth + frame signing verified live against a hosted Buzz relay (frame accepted structurally; rejection semantics above are relay-policy, exactly matching the linked handler).

### DIFF
--- a/plugins/platforms/buzz/adapter.py
+++ b/plugins/platforms/buzz/adapter.py
@@ -37,6 +37,7 @@ environment and is never logged.
 """
 
 import asyncio
+import hashlib
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Adds a substantial Buzz presence/telemetry feature to the adapter: kind:20001 presence (online/offline + 60s heartbeat), kind:20002 typing indicators with a 5s keepalive loop, auto-removed 👀 "seen" reactions, and NIP-44-encrypted kind:24200 observer frames, plus a new `plugins/platforms/buzz/nip44.py` NIP-44 v2 implementation. Thoughtfully documented and follows the native buzz-acp harness's cadence. The main concern is the ephemeral publish path described below.

**What it does**
- Presence: publishes "online" on connect, a 60s heartbeat loop, and "offline" on disconnect; relay TTL is 180s.
- Typing: throttled to one kind:20002 per channel per 3s; `send_typing` starts a 5s keepalive loop that re-publishes while a status phrase or an in-flight turn marker exists, and stops on reply.
- Seen: adds 👀 reaction on inbound dispatch and removes it once `send()` lands the reply (single slot per chat).
- Observer: kind:24200 frames NIP-44-encrypted to the owner's pubkey (from `BUZZ_OWNER_PUBKEY` or `BUZZ_AUTH_TAG`), routed via a dedicated ephemeral WS.
- `nip44.py`: ports NIP-44 v2 (ECDH x-coord → HKDF-SHA256 salt "nip44-v2" → ChaCha20 + HMAC, versioned payload), with a decrypt path for round-trip tests.

**Non-blocking notes**
- **Ephemeral publish opens a brand-new WebSocket per event.** `_send_ephemeral`'s "reuse active WS" branch is effectively dead code — it detects `_ws_active` and then just falls through to a dedicated `websockets.connect` + full NIP-42 AUTH handshake for every publish. With presence every 60s, typing every 3s, and keepalive every 5s, that's a steady stream of new connections. Works, but it's the highest-leverage improvement: hold a reusable publish socket (with reconnect), or reuse the inbound WS via a lock/queue. Also worth confirming per-connection rate limits won't trip on typing cadence.
- `_resolve_owner_pubkey`: if `BUZZ_OWNER_PUBKEY` is set but malformed (not 64 hex), it returns `None` and never falls back to `BUZZ_AUTH_TAG` — the "else the auth tag" fallback is unreachable for that case. Consider falling through on invalid length rather than returning early.
- `_pending_eyes` is one slot per chat: if several messages arrive before a reply, only the most recent 👀 is tracked and removed; earlier ones can linger. A small dict (message_id → task) or per-message tracking would avoid stale reactions.
- `nip44_decrypt` decrypts before verifying the HMAC — for a test/verification-only path this is fine, but flip the order (verify MAC, then decrypt) if it ever validates untrusted input.
- No test coverage visible in the diff for the new NIP-44 round-trip or the keepalive loop — a `nip44_encrypt`/`nip44_decrypt` round-trip test plus a heartbeat cadence test would meaningfully de-risk this crypto surface.

Issue #84718, opened 2026-08-12
Compaction keeps the imperative and drops the policy: skills pruned mid-task while todos are re-injected verbatim (v0.19.1, gpt-5.6-luna)


Thiago-Mota-Santos:
## Summary

Compaction preserves the **imperative** (the todo list) but destroys the **policy and the justification** (skill instructions, and the reasoning that produced each todo). The result is an agent that keeps executing a task item long after the context that would let it question that item is gone.

I have a full forensic trace of a single request — "the Hodle logo looks wrong, please use `h-logo` from the other repo" — that burned **2 sessions, 372 messages, 253 tool calls, 78 API calls and ~9.5M cache-read tokens** and ended with the agent working toward deleting a pre-existing, unrelated checkout route. The actual code change it needed was 8 files × 4 lines.

This is offered as the repro that #63278 ("Hermes is forgetful?!", `needs-repro`) is missing, and it is adjacent to #32156. It is a distinct defect from #78645 (which is a refactor of the same file).

## Environment

| | |
|---|---|
| Hermes Agent | v0.19.1 (2026.7.30), upstream `890e3c34`, install method `git` |
| Model (main loop) | `gpt-5.6-luna` via `openai-codex` (`https://chatgpt.com/backend-api/codex`) |
| Model (compression aux) | `gpt-5.6-luna`, same provider (`auxiliary.compression`) |
| Surface | Discord thread (`source=discord`, `chat_type=thread`), `require_mention: true`, `auto_thread: true` |
| Workspace | plain repo checkout (no worktree, no kanban card), dirty tree on a feature branch |
| OS | macOS (darwin 25.6.0) |

Relevant `config.yaml`:

```yaml
compression:
  enabled: true
  threshold: 0.5        # compress at 50% context fill
  target_ratio: 0.2     # squash to 20%
  protect_first_n: 3
  protect_last_n: 20
  in_place: false
  abort_on_summary_failure: false
skills:
  write_approval: false     # background review can rewrite SKILL.md unattended
curator:
  enabled: true
```

## Session metrics

| | Session 1 `20260812_152044_2d88bcc4` | Session 2 `20260812_154140_df4396` |
|---|---|---|
| Title | "Atualização do Logo Hodle no Frontend" | same, `#2` (auto-forked) |
| Messages | 203 | 175 (still growing at time of writing) |
| Tool calls | 142 | 111 |
| API calls | 47 | 31 |
| Input tokens | 390,859 | 251,578 |
[...]

PR #26981, opened 2026-05-16
Prevent todo snapshot injection after compression

YLChen-007:
## Summary

- Remove post-compression todo snapshot injection as a synthetic `user` message
- Keep todo state in the in-memory `TodoStore` instead of reintroducing raw todo text into compressed history
- Add regression coverage for instruction-like todo content during compression

Fixes #26979.

## Tests

- `scripts/run_tests.sh tests/run_agent/test_run_agent.py::test_compress_context_does_not_append_todo_snapshot_as_user_message tests/tools/test_todo_tool.py -q`

PR #97604, opened 2026-08-29
fix(compression): strip embedded stale todo snapshot from list message content

ericmaddox:
## What does this PR do?

Fixes a context compression token leak and stale-state accumulation bug in [`agent/conversation_compression.py`](file:///c:/Users/EricM/Dev/hermes-agent/agent/conversation_compression.py) where `_strip_stale_todo_snapshot()` failed to strip prior todo snapshots from list-structured (multimodal/block) user messages before re-injecting fresh state.

In Hermes Agent:
1. At every context compaction boundary, active todo items and pruned skill reload notices are merged onto the tail user turn.
2. `_strip_stale_todo_snapshot()` is intended to strip prior todo snapshots from the tail user message before re-injecting fresh state (#26981, #84718).
3. While string content was stripped via `content.find(TODO_INJECTION_HEADER)`, list content was filtered using `part["text"].startswith(TODO_INJECTION_HEADER)`.
4. When todo state was merged into an existing user text block (e.g. `[{"type": "text", "text": "user message\n\n### Current Todo List..."}, ...]`), `.startswith()` evaluated to `False`. The stale todo snapshot was never stripped, and the fresh todo block was appended again.
5. Over multi-hour agent sessions, repeated compaction passes accumulated duplicate stale todo snapshots and stale skill notices, ballooning token usage and corrupting the agent's task state.

The fix updates `_strip_stale_todo_snapshot()` for list content to search each text block for `TODO_INJECTION_HEADER` and truncate it via `text[:idx].rstrip()`, cleanly removing stale todo snapshots while preserving the user's prompt text and non-text parts.

## Related Issue

N/A

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `agent/conversation_compression.py`: Updated `_strip_stale_todo_snapshot()` to inspect text parts of list content with `text.find(TODO_INJECTION_HEADER)` and truncate to `text[:idx].rstrip()`.
[...]

PR #97613, opened 2026-08-29
fix(compression): strip embedded stale todo snapshot from list message content

lEWFkRAD:
## What does this PR do?

Fixes a silent stale-snapshot accumulation at context-compaction boundaries for **multimodal** trailing user turns.

`_strip_stale_todo_snapshot()` (`agent/conversation_compression.py`) is supposed to remove a previously merged todo snapshot before re-injecting a fresh one at each compaction boundary (#26981). The string-content path does this with `find()` + cut. The **list-content** path instead filtered parts whose text `startswith(TODO_INJECTION_HEADER)`.

But the injection site (`compress_context`) merges the snapshot onto the trailing user turn via `_append_text_to_content()`, which **appends after the existing text** (`context_compressor.py`: `[*content, text_block]` / string concat). So on a multimodal tail — content is a list because the turn carries an image — the stale header sits *mid-part*, and the `startswith` filter never matches. The stale snapshot survives; the fresh one is appended again. Repeated boundaries accumulate duplicate todo snapshots (and duplicate pruned-skill reload notices, #84718) inside every later text block.

The fix mirrors the string-path semantics in the list path: locate the header anywhere inside a text part and cut there; drop a text part that is nothing but snapshot scaffolding (so no empty `{"type":"text","text":""}` part is left behind). Non-text parts (image/audio) pass through untouched.

## Related Issue

No open issue for the list-content variant (the existing coverage #26981 / #84718 pin only the string path). Noticed while auditing the strip lifecycle against the injection site. Happy to open an issue first if maintainers prefer that flow.

## Type of Change

## Changes Made

- `agent/conversation_compression.py` — `_strip_stale_todo_snapshot()`: list-content branch now uses `find()` on each text part and cuts at the header (matching the string path), dropping whole-snapshot parts entirely.
- `tests/agent/test_skill_todo_retention_parity.py` — two regression tests in `TestNoticeStripLifecycle`: embedded-after-user-text snapshot stripped from a multimodal tail (image part preserved), and snapshot-only block still dropped entirely.

## How to Test

1. On plain `main`: `pytest tests/agent/test_skill_todo_retention_parity.py -q -k list_content` → `test_strip_removes_embedded_snapshot_from_list_content` **fails** (stale snapshot survives in the text part) — verified against `ac6c8028e`: 1 failed, 1 passed.
2. On this branch: `pytest tests/agent/test_skill_todo_retention_parity.py -q` → 14 passed (verified on Windows 11, Python 3.11.15).
3. Compression-adjacent regression batch: `pytest tests/agent/test_compression_adoption_preserves_live_tail.py tests/agent/test_compression_attempt_telemetry.py tests/agent/test_compression_count_warning_36908.py tests/agent/test_api_content_sidecar.py -q` → 36 passed.

### Code

- [x] I've run `pytest tests/ -q` and all tests pass — ran the targeted + compression-adjacent files from the changed subsystem; full suite pending in CI

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (docstring in the changed function updated)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (pure stdlib list/string logic, no platform constructs)

## Screenshots / Logs

Baseline proof (plain `ac6c8028e` + new tests): the failing assertion shows the text part retaining `[Your active task list was preserved across context compression]…[Skills pruned during compression…]…` after the strip; with the fix the part is exactly `"real user words"` and the image part is preserved.

### DIFF
--- a/agent/conversation_compression.py
+++ b/agent/conversation_compression.py
@@ -2428,17 +2428,33 @@ def _strip_stale_todo_snapshot(content: Any) -> Any:
             return content
         return content[:idx].rstrip()
     if isinstance(content, list):
-        return [
-            part
-            for part in content
-            if not (
-                isinstance(part, dict)
-                and part.get("type") == "text"
-                and str(part.get("text") or "")
-                .lstrip()
-                .startswith(TODO_INJECTION_HEADER)
-            )
-        ]
+        cleaned = []
+        for part in content:
+            if not isinstance(part, dict):
+                cleaned.append(part)
+                continue
+            if part.get("type") == "text":
+                # Mirror the string path: the snapshot is APPENDED after the
+                # user text by ``_append_text_to_content`` (see the injection
+                # site in ``compress_context``), so on multimodal tails the
+                # header sits mid-block and a startswith filter never matches.
+                # ``find`` + cut strips the embedded block; a block that is
+                # nothing but the snapshot drops entirely (no empty text part
+                # left behind).
+                text = str(part.get("text") or "")
+                idx = text.find(TODO_INJECTION_HEADER)
+                if idx == -1:
+                    cleaned.append(part)
+                    continue
+                stripped = text[:idx].rstrip()
+                if stripped:
+                    new_part = dict(part)
+                    new_part["text"] = stripped
+                    cleaned.append(new_part)
+                # else: whole part was snapshot scaffolding — drop it.
+            else:
+                cleaned.append(part)
+        return cleaned
     return content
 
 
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #97604. Both change the multimodal list-content stale-todo cleanup from a startswith filter to find-and-cut handling for a snapshot embedded after user text; #97604 is the earlier canonical repair.

lEWFkRAD:
Closing this draft in favor of the earlier canonical #97604. The claimed production duplication was not reproduced through the real append -> strip -> append flow on the current base: _append_text_to_content appends a separate {'type': 'text'} block for list content, and the replay retained a single TODO snapshot header. This test instead manually constructs a TODO header in the middle of an existing text part, so it does not establish that the production writer creates the reported state. Preserving the branch for reference; any follow-up should first demonstrate the state via the actual producer path on #97604.

PR #97615, opened 2026-08-29
feat(gateway): opt-in reaction-based busy acks (busy_ack_reaction)

robottwo:
## Summary
- New opt-in display setting `display.platforms.<platform>.busy_ack_reaction` (default `false`): when enabled and the platform adapter implements `send_reaction()`, the busy-input acknowledgment becomes a mode-aware emoji **reaction on the user's message** instead of a text bubble.
- Mode mapping: redirect / steer / queue / interrupt each get a distinct emoji, so the user can tell at a glance how their message was handled while the agent is mid-turn.
- Strict fallback: unsupported adapter, missing message id, a reaction that returns falsy, or any exception falls back to the existing text-bubble ack. Behavior when the setting is off is byte-identical to today.

## Why
On reaction-capable platforms (Buzz, Discord, Slack), a text bubble per busy ack clutters the timeline — the info fits naturally in a reaction, and reaction read-receipt UX already exists there.

## Implementation
- `gateway/run.py`: after the busy-ack timestamp latch, resolve `busy_ack_reaction` through the standard `resolve_display_setting` platform-scoped path; on success, short-circuit the text ack.
- `tests/gateway/test_busy_ack_reaction.py`: 4 regression tests — reaction ack when enabled, text fallback on missing `send_reaction`, fallback when the reaction send fails/returns false, and no-op when the setting is disabled.

## Verification
- `venv/bin/python -m pytest tests/gateway/test_busy_ack_reaction.py -o addopts= -q` — 4 passed

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -10745,6 +10745,51 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
 
         self._session_state(session_key).turn.busy_ack_ts = now
 
+        # Reaction-first busy ack: when the adapter supports send_reaction
+        # and the platform opt-in is enabled, acknowledge with a mode-aware
+        # emoji reaction on the user's message instead of a text bubble.
+        # Falls back to the text bubble when the reaction cannot be
+        # delivered (unsupported adapter, missing message id, send failure).
+        _BUSY_ACK_REACTIONS = {
+            "redirect": "↪️",
+            "steer": "⏩",
+            "queue": "⏳",
+            "interrupt": "⚡",
+        }
+        _ack_mode = (
+            "steer" if is_steer_mode
+            else "redirect" if is_redirect_mode
+            else "queue" if is_queue_mode
+            else "interrupt"
+        )
+        _reaction_ack_enabled = bool(
+            resolve_display_setting(
+                _load_gateway_config(),
+                platform_key,
+                "busy_ack_reaction",
+                False,
+            )
+        )
+        if (
+            _reaction_ack_enabled
+            and event.message_id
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Opt-in feature — when `display.platforms.<platform>.busy_ack_reaction` is enabled and the adapter supports `send_reaction()`, busy-input acknowledgments become mode-aware emoji reactions on the user's message instead of a text bubble. Clean fallback chain (unsupported / no message_id / send failure → text bubble), with solid async regression tests.

**What it does**
- In `_handle_active_session_busy_message`, computes `_ack_mode` (steer/redirect/queue/interrupt) and the emoji map, checks the opt-in via `resolve_display_setting`, and when enabled calls `adapter.send_reaction(chat_id, message_id, emoji)`.
- Returns `True` early on a successful reaction; otherwise falls through to the existing text-bubble path.
- New test file covers: reaction used when enabled (and no text bubble follows), fallback to text when the adapter lacks `send_reaction`, skip when disabled, and fallback on send failure.

**Non-blocking notes**
- The early `return True` on a successful reaction skips *everything* below in the handler — verify that nothing after the bubble-build block is required bookkeeping (logging/metrics/state) that should also run for the reaction path. `busy_ack_ts` is set before this block, so that part is covered, but worth confirming there's no ack telemetry further down.
- The `_BUSY_ACK_REACTIONS` dict and `_ack_mode` are computed on every busy message even when the reaction path is off — trivial, but could be module-level constants + a small helper for readability.
- `resolve_display_setting(_load_gateway_config(), ...)` reloads the gateway config on each busy ack; busy acks are infrequent, so acceptable, but a cached config would avoid the repeated parse.

PR #97617, opened 2026-08-29
fix(gateway): label busy-path steer/redirect follow-ups with the shared-session sender

### LINKED ISSUE #97569: Mid-turn messages bypass sender labelling when busy_input_mode is interrupt or steer
### Summary

When `display.busy_input_mode` is `interrupt` (the default) or `steer`, a message that arrives
**while a turn is already running** bypasses `_prepare_inbound_message_text()` entirely. It is
delivered to the agent as a bare string, losing the sender label, the reply context, the message id,
the timestamp and any forwarding information.

In a multi-user chat this is not cosmetic: the agent cannot tell who spoke, and any downstream
automation keyed on the sender prefix silently stops firing. Nothing errors, and nothing in the
transcript indicates a message was degraded rather than absent.

Observed on 0.19.1 (`v2026.7.30`), Telegram adapter, group chat.

### Where it happens

The busy-message handler tests for a redirectable event, then passes only the text:

- `gateway/run.py:10144-10155` — the guard requires `MessageType.TEXT` with no `media_urls` and no
  `media_types`, then calls `redirect()` with `(event.text or "").strip()`. The `event` itself, which
  carries everything else, is not passed on.
- `agent/conversation_loop.py:306-337` — that bare string is persisted as the user content, so the
  loss is durable, not just in-flight.
- `gateway/run.py:10111-10143` — the `steer` path does the same via `running_agent.steer()`.

The normal path does considerably more:

- `gateway/run.py:17873-17897` — adds the sender prefix.
- `gateway/run.py:18121-18135` — adds reply context.
- `gateway/run.py:19840-19868` — persists the timestamp.

The Telegram adapter does build the reply context
(`plugins/platforms/telegram/adapter.py:10479-10536`); it simply never reaches `redirect()`.

### How to reproduce

1. Set `display.busy_input_mode: interrupt` (the default) with more than one person in a chat.
2. Send a message that starts a turn.
3. While the agent is still working, have a **second** person send a plain text message.
4. Inspect the stored conversation. The second message is present with no sender prefix, while
   messages sent between turns in the same session carry theirs.

A discriminator over the stored rows: mid-turn arrivals are the ones whose API content contains
`interrupted`. In one session that separated the two populations cleanly — every unlabelled message
matched, every labelled one did not.

### Why `queue` is not a complete answer

`queue` does avoid it: the whole event stays in the FIFO (`gateway/run.py:10175-10176`) and at end of
turn re-enters `_prepare_inbound_message_text()` (`gateway/run.py:29182-29192`), so the label is
applied. But it is not the default, it changes the interaction model (the agent can no longer be
interrupted), and it does not help `interrupt` or `steer` users.

### Suggested fix

Pass the `event` to `redirect()`/`steer()` rather than a bare string, and run the same preprocessing
the queued path runs — or, more narrowly, apply `_prepare_inbound_message_text()` to the text before
handing it over. The labelling logic already exists and is already reached by one of the three modes;
this is about the other two calling it.

### One adjacent papercut

`display.busy_input_mode` recognises exactly `interrupt`, `queue` and `steer`. Any other value
silently becomes `interrupt` (`gateway/run.py:9344-9354`). A typo in the config therefore selects the
defective behaviour with no warning. A log line naming the resolved mode at startup, or a warning on
an unrecognised value, would make that visible — the neighbouring `service_tier` loader already logs
`Unknown service_tier '%s', ignoring`.

liuhao1024:
## What does this PR do?

Mid-turn follow-ups delivered while a turn is already running (`display.busy_input_mode: interrupt` or `steer`) bypassed `_prepare_inbound_message_text()` entirely — the busy handler passed only `(event.text or "").strip()` to `steer()`/`redirect()`. Two kinds of context were silently dropped:

1. **Sender attribution** — in shared multi-user sessions the agent could not tell who spoke, and the persisted transcript showed unlabelled mid-turn arrivals while between-turn messages carried theirs (#97569). `queue` mode was unaffected because the whole event stays in the FIFO and re-enters the normal pipeline.
2. **Reply-to disambiguation** — a reply-quote arriving mid-turn reached the agent as bare text with no `[Replying to: ...]` pointer, so the agent had to guess which prior message was being referenced — exactly when several drafts/answers are typically in flight. Reported independently with a real incident in #101866 (triaged as a duplicate of #97569): a mid-turn "yes, send the same" reply-quote lost its pointer and the agent resolved it against the wrong draft.

The first commit extracts the normal inbound path's sender-attribution block (shared-session detection, untrusted-name neutralization, Slack `<@U...>` enrichment) into `_label_shared_session_sender_text()` and routes both busy-path payloads through it. The second commit extracts the `[Replying to: ...]` prefix builder into a module-level helper and applies it in `_prepare_busy_steer_text()` (non-empty payloads only, so the empty-steer fallback to queue semantics is preserved) and around the labeled redirect payload — matching the normal inbound path's ordering (reply-to pointer outermost, sender label inside). Steered injections are also logged like normal inbound messages, since the busy path skips that pipeline's log line. DMs and per-user-isolated sessions label nothing — single-user behavior is byte-identical to before.

Scope note: this PR fixes sender attribution and reply context. The issue also mentions message-id/timestamp persistence being skipped on the busy path; that is intentionally left out to keep this diff minimal.

## Related Issue

Fixes #97569

## Type of Change

## Changes Made

- `gateway/run.py` — added `GatewayRunner._label_shared_session_sender_text()`: the shared-session sender-label primitive (was an inline block in `_prepare_inbound_message_text`)
- `gateway/run.py` — added module-level `_build_reply_to_disambiguation_prefix()`: the `[Replying to: ...]` pointer builder (was an inline block in `_prepare_inbound_message_text`)
- `gateway/run.py` — `_prepare_inbound_message_text()` now calls both helpers instead of the inline blocks (no behavior change on this path)
- `gateway/run.py` — busy steer path: `_prepare_busy_steer_text()` now applies the sender label and the reply-to prefix itself (empty payloads still return empty, preserving the queue-fallback semantics) and logs the injection
- `gateway/run.py` — busy interrupt/redirect path: `running_agent.redirect()` payload now carries the sender label with the reply-to prefix wrapped outside it
- `tests/gateway/test_busy_sender_labeling_97569.py` — new: steer/redirect labeling in shared groups, reply-to + label ordering on both busy paths, DM/isolated-group no-label behavior pins, hostile-display-name neutralization, Slack user-id enrichment, normal-path labeling preserved
- `tests/gateway/test_reply_to_injection.py` — new: own-message prefix variant on the normal path, busy-steer prefix injection, plain-steer verbatim pin, empty-payload stays empty

## How to Test

1. `uv run --extra dev pytest tests/gateway/test_busy_sender_labeling_97569.py tests/gateway/test_reply_to_injection.py -q`
   - Observed result: 15 passed.
2. `uv run --extra dev pytest tests/gateway/test_busy_session_ack.py tests/test_tui_gateway_queue_on_busy.py tests/run_agent/test_steer.py tests/gateway/test_busy_session_auth_bypass.py tests/gateway/test_multiplex_busy_input_mode.py tests/gateway/test_steer_fifo_overwrite.py tests/gateway/test_steer_command.py -q`
   - Observed result: 97 passed — existing busy-path behavior (including the unlabeled single-user payloads) is unchanged.
3. `uv run --extra dev pytest tests/gateway/test_shared_group_sender_prefix.py tests/gateway/test_reply_to_injection.py tests/gateway/test_busy_sender_labeling_97569.py tests/gateway/test_document_context_note.py -q`
   - Observed result: 23 passed — normal-path labeling and reply injection preserved.
4. Manual repro of the incident shape: Telegram DM, `busy_input_mode: steer`, start a long turn, reply-quote an earlier bot message with "yes, send this" while it runs. Should pass: the steered payload in `agent.log` starts with `[Replying to your previous message: "..."]`, and gateway.log shows the `inbound message (busy-steer): ... reply_to_id=...` line.

### Code

## Screenshots / Logs

```
$ uv run --extra dev pytest tests/gateway/test_busy_sender_labeling_97569.py tests/gateway/test_reply_to_injection.py -q
15 passed in 0.83s
```


Enough1122:

**Overall:** Fixes #97569 — mid-turn `steer()`/`redirect()` follow-ups in shared multi-user sessions bypassed `_prepare_inbound_message_text`, so busy-path payloads reached the agent/transcript with no sender attribution. The labeling logic is extracted into one shared `_label_shared_session_sender_text` method used by both the normal inbound path and the busy path, so untrusted-name neutralization and Slack user-id enrichment can't drift. Clean refactor with good regression coverage.

**What it does**
- New `_label_shared_session_sender_text(text, source)`: returns text unchanged unless it's a shared multi-user session with `source.user_name`, then prefixes `[{neutralized name}] ` (and Slack's verifiable `<@U...>` id) — exactly the logic previously inlined in `_prepare_inbound_message_text`, now with the `group_sessions_per_user` / `thread_sessions_per_user` reads moved inside.
- Busy `steer()` and `redirect()` calls now route their payload through the helper before handing to the running agent.
- `_prepare_inbound_message_text` delegates to the same helper, preserving existing behavior.
- 7 tests: steer/redirect labeling, DM and per-user-isolated sessions stay unlabeled (byte-identical pin), hostile display-name neutralization, Slack `<@U...>` enrichment, and normal-path parity.

**Non-blocking notes**
- Both busy branches now call the helper, which re-reads `self.config` attributes and re-derives `is_shared_multi_user_session(...)` on every mid-turn message — negligible cost, and keeps behavior identical to the previous inline computation.
- The helper defaults (`group_sessions_per_user=True`, `thread_sessions_per_user=False`) are duplicated in the helper body via `getattr(self.config, ..., default)`; the values mirror what `_prepare_inbound_message_text` used, so no drift — just noting they're now spread across two spots (the helper and `build_session_key`'s expectations) should they ever need to change.
- The unused `_thread_sessions_per_user` local removal in `_prepare_inbound_message_text` is a nice cleanup; the eager-labels-then-enriches ordering (prefix applied before channel-context backfill prepend) is preserved from before.

liuhao1024:
Thanks for the careful read — the extraction exists precisely so the normal and busy paths can't drift apart, and the non-blocking notes match the intent (behavior-preserving refactor, negligible per-message cost).

Issue #94724, opened 2026-08-25
Tracking: Desktop persistent multi-gateway connections — CAMPAIGN COMPLETE (29 PRs)


teknium1:
# Tracking: Desktop persistent multi-gateway connections — maintenance: externally-blocked + decisions only

**Status: campaign complete — final salvage wave merged, close-out done.** **29 PRs merged** across the campaign, 2 same-day regressions fixed, all salvage clusters shipped. The 15 salvaged originals and 6 resolved issues are closed with per-author credit; what remains below is **only** maintainer decisions and externally-blocked items. This body is the living index; comments carry the historical wave-by-wave detail.

## Shipped (all rebase-merged, contributor authorship preserved)

**Wave 1 — class-consolidated salvages (Aug 26):**
| PR | SHA | Scope |
|---|---|---|
| #95080 | 682a34a7c | Class 1 route identity |
| #95081 | 71d804d00 | Class 2 owning route |
| #95082 | b52b05c1d | Class 3 secondary lifecycle |
| #95083 | dce4abe91 | Class 3 gateway startup reconnect (py) |
| #95084 | e0210ab6c | Class 4 reconnect integrity |
| #95085 | 25d46c788 | Classes 5+6 auth envelope + process lifecycle |
| #95087 | 5a285d343 | Classes 7+8 scoping + session truth |
| #95088 | 266459964 | Routed history / transport auth (py) |
| #95073 | afee35700 | Subagent notification suppression (campaign infra) |

**Wave 2 (Aug 26):**
| PR | SHA | Scope |
|---|---|---|
| #95407 | 8e9459c97 | Phase A: durable session ownership (connection_id on rows, bounded boot waits) |
| #95495 | b455abe0b | Phase B: transport (bounded waits, 4001 poll-storm stop, warm-transcript release, #95393 switcher repaint) |
| #95496 | 600d5166f | Phase B: delivery-ledger reconnect sweep (py) |
| #95497 | bb3421bf2 | Phase B: sleep/wake SSH revalidation + single-owner dial claim |
| #95634 | 9aa7530f7 | Phase C: per-connection cookie auth, registry per-entry quarantine, no ghost loopback spawns, MCP reconnect classification, SSH quit escalation |
| #95635 | fd565c80e | Phase D: confirm_required handshake, instant bot-chat paint (#89843), fleet profile rail (#92384) |
| #95086 | b3a2065ff | Windows updater scoped kill (Windows-runner A/B proof in PR body) |

[...]

PR #84803, opened 2026-08-12
fix(gateway): reaffirm YAML skins on connect via skin.changed

Zirillian:
## What does this PR do?

Fixes custom YAML skins (`~/.hermes/skins/*.yaml`, e.g. mythos) resetting to **nous** after every Desktop restart, **without requiring a Desktop client rebuild**.

`gateway.ready` seeds the skin with `apply: false` so a reconnect cannot stomp a built-in Appearance pick. That same rule leaves user YAML skins stuck: `normalizeSkin` rejects the persisted name until the seed lands, then never applies it. Built-ins already persist because they resolve on the first frame.

This PR sends a follow-up `skin.changed` **only** when the resolved skin is an on-disk user YAML (not a Desktop built-in). Current Desktop clients already treat seed-only + same-name `skin.changed` as missed-activation recovery (see `backend-sync` tests / #69533).

This is **not** another renderer-only persist PR. It is the gateway half that makes **already-shipped** Desktop binaries recover. Renderer-only PRs (#71447, #76648, #82483, #82951, #74018) still need a new Desktop build to help remote/Windows clients.

Verified on Windows Desktop (official binary) + Linux isolated `hermes serve --isolated`: reopen paints nous briefly, then auto-switches to the YAML skin with no Appearance click.

## Related Issue

Fixes #71446

Also related (same class, first-use / renderer): #73987

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `tui_gateway/server.py` — `should_reaffirm_yaml_skin_on_connect()`; true only for on-disk YAML names that are not Desktop built-ins (`nous`, `midnight`, `ember`, `mono`, `cyberpunk`, `slate`, `default`)
- `tui_gateway/ws.py` — after a successful `gateway.ready` seed + `register_live_transport`, emit `skin.changed` when the helper is true
- `tests/tui_gateway/test_protocol.py` — helper: mythos True; nous/slate False; missing file False; `{}` / `None` False

## How to Test

1. `hermes config set display.skin mythos` (or any name with `~/.hermes/skins/<name>.yaml`)
2. Fully quit Desktop (tray → Quit), recycle leftover `hermes serve --isolated` if needed, reopen and reconnect
3. **Before:** stays on nous until you re-pick
4. **After:** brief nous (cold-boot fallback), then auto-switch to the YAML skin. No Appearance click.
5. Pick a built-in (slate) in Appearance, restart — still slate. This PR must not reaffirm built-ins.

[...]

Issue #83134, opened 2026-08-10
[Bug]: Remote Desktop reconnect storm drives dashboard to ~190% CPU


AgatElite:
## Bug Description

After updating Hermes Desktop to v0.20.0, connecting from my MacBook to a remote Hermes server causes a continuous WebSocket/API reconnect loop.

Desktop shows:

> Runtime not ready. Hermes desktop cannot verify the running back end on startup.

The composer placeholder continuously changes, although existing sessions remain visible. While Desktop is open, the remote dashboard process reaches approximately 173–190% CPU. Quitting Desktop immediately returns it to 0–1%.

The server is reachable throughout: Tailscale is connected and `/api/health` returns HTTP 200.

## Steps to Reproduce

1. Run Hermes v0.20.0 on a remote Ubuntu server.
2. Confirm the dashboard is healthy and `/api/health` returns 200.
3. Open Hermes Desktop v0.20.0 on macOS using the saved remote Tailscale HTTPS URL.
4. Observe the runtime warning, changing composer placeholder, and server CPU.
5. Quit Desktop; server load immediately returns to idle.

Restarting only `hermes-dashboard.service` and repeating these steps produces the same result.

## Expected Behavior

Desktop establishes one stable WebSocket connection to the healthy remote backend without flooding it with reconnects or API requests.

## Actual Behavior

A controlled reproduction after a clean dashboard restart recorded:

- 55 WebSocket opens and 56 closes in 12 seconds
- 564 API GET requests in 12 seconds
- Dashboard CPU at 173–190%
- Dashboard CPU returning to 0–1% after quitting Desktop

The server repeatedly accepts a WebSocket before the client disconnects:

    ws accepted peer=127.0.0.1:...
    ws send failed ... error_type=WebSocketDisconnect
    RuntimeError: WebSocket is not connected. Need to call "accept" first.

Connection summaries report `parse_errors=0` and `dispatch_crashes=0`.

## Affected Component

Hermes Desktop remote connection / TUI gateway and dashboard WebSocket path.

Messaging platform: N/A.

## Debug Report

Credential-redacted links, expiring after six hours:

- Report: https://dpaste.com/5L3SZ6GUT
- Focused GUI log: https://dpaste.com/9MSCQETQX

## Environment

- Desktop: macOS 26.5.2, Apple Silicon
- Hermes Desktop: v0.20.0 (+7), commit `4e46129ff`
- Remote host: Ubuntu 24.04.4, Linux x86_64
[...]

PR #73821, opened 2026-07-29
fix(desktop): settle the gateway connect on gateway.ready and keep a stale boot from finishing after a switch

Sora-bluesky:
## What does this PR do?

Reworked on current main (the earlier version of this branch retried the boot-time WS dial with its own ladder; upstream 2b7f49673f now owns boot retry policy, so that part is gone).

`JsonRpcGatewayClient.connect()` resolved on the raw WebSocket `open` and `connectionState` became `'open'` at the same time. The desktop boot in `use-gateway-boot.ts` treats a resolved `connect()` as a usable gateway (adopts the profile, hydrates config/sessions, calls `completeDesktopBoot()`), and its reconnect bookkeeping (`reconnectAttempt`, `reconnectFailingSince`, `escalated`) resets on every `'open'`. So a gateway that accepts the upgrade and closes right away looks healthy: boot can complete on a dead socket, and a persistently rejected socket never reaches the reconnect escalation (5 minutes since 8db59d08f4) because each `'open'` clears the failure clock. `request()` only checked the socket's `readyState`, so callers could send during the handshake. This is the client half of #88607: once `/api/ws` rejections are delivered as close codes (as #88619 does for the other endpoints), the client has to treat `open` followed by a close as a rejection, not a connection.

Every `/api/ws` connection starts with a `gateway.ready` event (`tui_gateway/ws.py::handle_ws` writes it right after `accept()`), so that frame is where the connection is known to be usable.

- `connect()` resolves, and the state becomes `'open'`, only on the first frame, which must be `gateway.ready`. A different first frame is a protocol failure (state `'error'`, generic connect error). Raw `open` leaves the state `'connecting'` and the promise pending. `request()` requires the semantic `'open'`.
- A close during the handshake settles the attempt immediately with a `GatewayConnectError` (exported from `@hermes/shared`) carrying `wsCloseCode`. 4401 sets `needsOauthLogin` and uses the new `authRejectedErrorMessage` option, so `isGatewayReauthRequired()` is true and the boot-failure overlay routes to the sign-in path (`HermesGateway` passes the canonical "Your remote gateway session has expired…" text that `isRemoteReauthError()` matches). 4403/4400 keep the code without the auth marker. `onSocketClose` still applies to established connections only.
[...]

PR #97624, opened 2026-08-29
fix(gateway): send WebSocket readiness without executor wait

aryaxo:
## Summary

- send `gateway.ready` from a thread-safe, last-good skin snapshot instead of waiting on the shared executor
- prime the skin snapshot before the dashboard starts accepting WebSocket connections
- collapse cold-cache fallback work into one process-wide background refresh so reconnect bursts cannot queue one resolver per peer
- serialize process-global skin resolution and revision-guard late refreshes so stale workers cannot overwrite or broadcast newer state
- bind fallback refresh task reuse to its owning event loop and notify connected cold-cache clients with `skin.changed` when resolution completes

## Problem

Each WebSocket handshake currently awaits `asyncio.to_thread(server.resolve_skin)` before emitting `gateway.ready`. Under executor/GIL pressure, the resolver itself may be fast once scheduled, but the wait can exceed the Desktop readiness timeout. The client reconnects, each reconnect queues another resolver, and the resulting feedback loop can leave an otherwise healthy dashboard appearing unavailable while unrelated HTTP routes remain responsive.

This is not deployment-specific: any Desktop or other WebSocket client can hit it when the dashboard is busy.

## Compatibility

A normally started dashboard primes the cache synchronously during existing module warm-up. Standalone/test entry points that somehow connect with a cold cache receive readiness immediately with an empty skin payload, allowing clients to retain their persisted theme. One background refresh repopulates the cache and emits `skin.changed` to the already-registered clients when the resolved snapshot is still current. Failed refreshes preserve the last-good snapshot.

## Related work

- Follow-up to #60800 / #77814: that work moved skin resolution off the event loop; this PR removes it from the WebSocket readiness critical path when the shared executor is saturated.
- Related to #83134 and tracker #94724: those cover broader reconnect-storm and retry/backoff failures. This PR fixes one server-side amplification trigger and does not replace client-side reconnect/backoff work.
- Compatible with #73821 (client readiness handshake) and #84803 (custom YAML skin reaffirmation).

## Test plan

- `uv run --with pytest python -m pytest -q tests/test_tui_gateway_ws.py tests/tui_gateway/test_cold_start_gil_stall.py tests/tui_gateway/test_94895_backend_heartbeat.py tests/tui_gateway/test_entry_picker_prewarm.py tests/tui_gateway/test_startup_orphan_sweep.py tests/tui_gateway/test_protocol.py tests/test_cli_skin_integration.py tests/cli/test_cli_skin_integration.py tests/hermes_cli/test_skin_engine.py tests/hermes_cli/test_skin_palettes.py tests/hermes_cli/test_skin_cmd.py` — **156 passed**
- Ruff on all five changed files — passed
- `compileall` on all five changed files — passed
- `git diff --check` — passed

Behavioral coverage includes readiness beating blocked resolution, reconnect-burst deduplication, late cold-client notification, last-good preservation on resolver failure, stale startup snapshot correction, stale revision rejection, and cross-event-loop task ownership.

## Manual verification

On Windows, after restarting the dashboard with this patch:

- 20 simultaneous authenticated WebSocket connections all received `gateway.ready` with a populated skin
- ready latency: **79 ms minimum, 131 ms median, 251 ms maximum**
- an older 4,216-message session resumed with `defer_history=true` and `omit_messages=true` in **741 ms**, with no message history included in the response

### DIFF
--- a/hermes_cli/web_server.py
+++ b/hermes_cli/web_server.py
@@ -326,8 +326,8 @@ def _warm_gateway_module() -> None:
     RPC burst (``setup.status``, ``setup.runtime_check``,
     ``gateway.ready``→``resolve_skin``) pull in several *other* heavy
     chains that were still imported on the loop thread, contributing to
-    the ~14s cold-start stall (#60800). Warm them all here so the cost
-    is paid in a worker thread while the server socket is already open.
+    the ~14s cold-start stall (#60800). Warm them all before the lifespan
+    yields so the first connection is never exposed to that import cost.
     """
     for mod in (
         "hermes_cli.gateway",
--- a/hermes_cli/web_server.py
+++ b/hermes_cli/web_server.py
@@ -352,6 +352,16 @@ def _warm_gateway_module() -> None:
         except Exception:
             pass
 
+    # Prime the no-I/O snapshot used by gateway.ready. This happens before the
+    # lifespan yields (and therefore before sockets are accepted), so a busy
+    # default executor can never hold the liveness frame hostage later.
+    try:
+        from tui_gateway import server as gateway_server
+
+        gateway_server.resolve_skin()
+    except Exception:
+        pass
+
 
 def _resolve_restart_drain_timeout() -> float:
     try:
--- a/hermes_cli/web_server.py
+++ b/hermes_cli/web_server.py
@@ -337,10 +337,9 @@ def _warm_gateway_module() -> None:
         "hermes_cli.auth",
         "hermes_cli.copilot_auth",
         "hermes_cli.runtime_provider",
-        # resolve_skin() reads config + initialises the skin engine.
-        # Even though handle_ws now calls it via asyncio.to_thread
-        # (see tui_gateway/ws.py), warming it here avoids the first-call
-        # import cost inside that thread.
+        # resolve_skin() reads config + initialises the skin engine. Standalone
+        # cold-cache paths still resolve it in a background thread; warming the
+        # module here avoids first-call import cost inside that fallback.
         "hermes_cli.skin_engine",
         # model.options / picker context — parses provider catalogs and
         # the models.dev cache on first use.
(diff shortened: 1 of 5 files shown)


Enough1122:

**Overall:** Makes WebSocket `gateway.ready` a true liveness signal that never waits on the shared default executor. `web_server` primes a no-I/O skin snapshot before accepting sockets; `handle_ws` reads that cached snapshot synchronously (empty-skin + one background refresh on cold cache), and `resolve_skin` refreshes the cache whenever it runs. Reconnect bursts collapse to a single process-wide refresh. Well-designed with strong behavioral tests.

**What it does**
- `tui_gateway/server.py`: new `_skin_payload_cache` + lock; `get_cached_skin_payload()` returns a lock-guarded copy; `resolve_skin` updates the cache on success.
- `tui_gateway/ws.py`: `handle_ws` reads `get_cached_skin_payload()` inline instead of `await asyncio.to_thread(server.resolve_skin)`; on a cold (empty) cache it starts one reused `_ensure_skin_cache_refresh()` task (single `_skin_refresh_task`, off-loop via `to_thread`).
- `hermes_cli/web_server.py`: `_warm_gateway_module` primes `resolve_skin()` before lifespan yields, so the first socket always has a snapshot.
- Tests cover: ready frame beating a blocked skin lookup, reconnect burst → exactly one resolver call, ready payload served from cache without calling the resolver, and warm-module priming.

**Non-blocking notes**
- `_skin_refresh_task` is a module-global created on whichever event loop first needs it; `handle_ws` does `_skin_refresh_task.done()` from potentially different loops (e.g., per-test `asyncio.run`). In production there's one loop so it's fine, and tests reset it — but a loop-strict guard (store loop + task together, or use a per-loop weak key) would harden against multi-loop embedding.
- `get_cached_skin_payload` returns a shallow copy (`dict(...)`), so nested `colors`/palette dicts are shared references with the cache. Since the payload is JSON-serialized for the frame, this is safe today, but worth a comment if anything downstream mutates the returned structure.
- If a warm prime failed at startup and the first connection hits a cold cache, the ready frame goes out with an empty skin and the refresh runs in background — clients keep their persisted theme, so this is acceptable, but there's no explicit retry/notification when the snapshot eventually lands (relies on the existing `skin.changed` broadcast path). Worth confirming that path covers the cold-start-late-prime case.

PR #97625, opened 2026-08-29
fix(state): skip archive_and_compact when the live transcript already matches

### LINKED ISSUE #97321: [Bug]: compression summary message can be inserted twice in a single cycle (race between preflight and tool-loop tail paths)
## Summary

On MiniMax-M3 (Anthropic wire, 1M context), `archive_and_compact()` can insert the **same compaction summary row twice** in a single compaction cycle, producing byte-identical duplicate `[CONTEXT COMPACTION — REFERENCE ONLY]` markers in the chat transcript. Markers are written 7 seconds apart but are otherwise identical — same content, same character count, same `compacted=1` flag. The effect is cosmetic but visible: chat scroll fills with redundant "[CONTEXT COMPACTION]" handoffs and `session_search` recall doubles for those rows.

## Reproduction (measured, real session)

Hermes v0.20.6, desktop, MiniMax-M3 on `api.minimax.io/anthropic`, model `MiniMax-M3`, session `20260828_115258_fa2ef8` (8 minutes, 49 API calls, TripForge dev).

| msg_id | role | length | compacted | within session clock |
|---|---|---|---|---|
| 428336 | assistant | 3552 | 1 | +0s |
| 428396 | assistant | **4425** | 1 | +40s |
| 428402 | assistant | 3814 | 1 | +42s |
| 428436 | assistant | **4425** | 1 | +74s |
| 428781 | assistant | 6085 | 1 | +372s |
| 428797 | assistant | **4425** | 1 | +381s |
| 428903 | assistant | 7531 | 1 | +526s |
| 428905 | assistant | **4425** | 1 | +527s |

Pairs at rows `428396/428436`, `428797/428905` (and one more I truncated) are **byte-for-byte identical** — same content text, same character count, written within the same second window. The interleaved rows of different length suggest the compactor was called twice in close succession and the second call re-persisted the first call's summary verbatim.

```
$ sqlite3 state.db "SELECT content FROM messages WHERE id IN (428396, 428436)"
# both rows return identical 4425-char content starting:
# [CONTEXT COMPACTION — REFERENCE ONLY] Earlier turns were compacted into the
# summary below. This is a handoff from a previous context window — treat it
# as background reference, NOT as active instructions. ... [tripforge dev
# context...]
```

## Expected behavior

Each successful `archive_and_compact()` call should insert the new summary exactly once. Subsequent calls should insert a new summary with a strictly later timestamp and ideally different content (the tail it summarizes has moved forward).

## Hypothesis

Two compression paths race during a single turn:
1. **Turn-prologue path**: `conversation_loop.py` checks preflight pressure and may fire compaction before the next API call.
2. **Tool-loop tail path**: after a tool returns, `agent_runtime_helpers.py` (or similar) checks if the new tool output pushed pressure over the threshold and may fire compaction mid-turn.

If both paths cross the threshold at the same wall-clock moment, both proceed to `_insert_message_rows(...)` for the summary. The `_persist_compaction` / `_record_summary_insert` helper (or whatever the canonical write site is) has **no `threading.Lock` / `asyncio.Lock`** — I grepped `agent/context_compressor.py` for `Lock`, `RLock`, `threading`, and `asyncio.Lock` and found zero matches in any insert path. The first path's INSERT lands, the second path's INSERT lands, and the DB now has two byte-identical summary rows for one logical compaction.

Suggested fix: either
- a coarse `threading.Lock` around the entire compaction cycle (acquire at `should_compress()` → release after `_insert_message_rows()` returns), or
- `INSERT ... ON CONFLICT DO NOTHING` keyed on `(session_id, sequence_or_role_marker)` where the role marker is a unique sentinel like `"__compression_summary__"`, or
- a per-turn gate that records `last_compaction_attempt_ts` and refuses to fire a second compaction within N seconds.

## Evidence: per-row timing suggests two writes within ~3-7 seconds of each other

The two pairs I quoted (428396/428436 = +34s, 428797/428905 = +9s) are not 7s apart but the row IDs differ by 40 and 8 respectively — small gaps consistent with both paths racing through `_persist_compaction` before either commits. The DB's auto-incrementing row id is the most reliable clock we have here since session messages are inserted in commit order.

## Environment

- Hermes Agent, `main` @ `fd3a783a3` (1 commit behind origin/main = `baa344dee`, unrelated)
- TripForge dev worktree at `/workspace/tripforge`
- Provider: `minimax`, model `MiniMax-M3`, base_url `https://api.minimax.io/anthropic`
- `compression.threshold: 0.55`, `proactive_prune_tokens: 100000`, `target_ratio: 0.3` (defaults before my 2026-08-28 stop-gap)
- Sibling finding: this is **distinct from #86366** (which covers `compacted=1` flagging the carried-forward tail verbatim, leaving 12.6% recall duplicates). That bug is about row labeling; this one is about summary-message insert race.
- Adjacent to #75145 (P3, transcript replay dedup at session_search level). Different mechanism.
- Adjacent to #58407 (P2, auto-lower every-turn loop). Different mechanism but same family: post-compaction state isn't propagated back to preflight estimates.

## Suggested next step

Add a `threading.Lock` (or `asyncio.Lock` if the compression path is async — the rest of the loop is mixed-mode) around the canonical `_persist_compaction` entry point. Verify by re-running my reproduction: 8 turns in 8 min should now produce 4 distinct summary rows (one per logical compaction), not 8 with 4 byte-identical pairs.

I can ship the patch + a regression test if helpful — happy to do that as a follow-up PR.

686f6c61:
## What does this PR do?

`archive_and_compact()` always soft-archived the live transcript and inserted the compacted set, even when that set was already the live transcript. A second compression path in the same cycle (turn preflight + tool-loop tail) therefore published a second identical `[CONTEXT COMPACTION — REFERENCE ONLY]` row. `session_search` includes compacted rows, so the duplicate was visible.

If the active messages already match the compacted set (same roles and contents, same order), the commit is now a no-op.

## Related Issue

Fixes #97321

## Type of Change

- [ ] Security fix
- [ ] Documentation update

## Changes Made

- `hermes_state.py`: `_active_messages_match_compacted` + early return in `archive_and_compact`
- `tests/test_compression_watermark_commit.py`: second compact keeps the same live ids and does not duplicate the summary among compacted rows

## How to Test

```text
python -m pytest tests/test_compression_watermark_commit.py -q
# 17 passed
```

### Code

- [x] I've run the targeted tests listed above
- [x] I've tested on my platform: Ubuntu (unit tests)

### DIFF
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -11823,6 +11823,20 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
                         "commit; refusing to publish a stale compaction"
                     )
 
+            if self._active_messages_match_compacted(
+                conn, session_id, compacted_messages
+            ):
+                # A second compression path in the same cycle (preflight +
+                # tool-loop tail) can republish the identical summary.
+                # archive-then-insert would duplicate the row as compacted=1
+                # and session_search would return both (#97321).
+                count_row = conn.execute(
+                    "SELECT COUNT(*) AS n FROM messages "
+                    "WHERE session_id = ? AND active = 1",
+                    (session_id,),
+                ).fetchone()
+                return int(count_row["n"] if count_row is not None else 0)
+
             patched_model_config = None
             if model_config_patch is not None:
                 # on_missing="raise": a prune/compaction must not commit
@@ -11905,6 +11919,27 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
             return inserted
 
         return self._execute_write(_do)
+
+    def _active_messages_match_compacted(
+        self,
+        conn,
+        session_id: str,
+        compacted_messages: List[Dict[str, Any]],
+    ) -> bool:
+        """True when the live transcript is already *compacted_messages*."""
+        rows = conn.execute(
+            "SELECT role, content FROM messages "
+            "WHERE session_id = ? AND active = 1 ORDER BY id",
+            (session_id,),
+        ).fetchall()
+        if len(rows) != len(compacted_messages):
+            return False
+        for row, msg in zip(rows, compacted_messages):
+            if (row["role"] or "") != (msg.get("role") or ""):
+                return False
+            if row["content"] != self._encode_content(msg.get("content")):
+                return False
+        return True
 
     def _message_column_names(self, conn) -> List[str]:
         """Column names of the messages table, cached per-connection era."""
--- a/tests/test_compression_watermark_commit.py
+++ b/tests/test_compression_watermark_commit.py
@@ -41,6 +41,31 @@ SUMMARY = [
 
 
 class TestWatermarkCommit:
+    def test_identical_second_compact_does_not_duplicate_summary(
+        self, db: SessionDB
+    ) -> None:
+        """Preflight + tool-loop must not insert the same summary twice (#97321)."""
+        _seed(db)
+        db.archive_and_compact("sess1", SUMMARY)
+        live_after_first = db.get_messages("sess1")
+        first_ids = [row["id"] for row in live_after_first]
+        assert [row["content"] for row in live_after_first] == [
+            SUMMARY[0]["content"],
+            SUMMARY[1]["content"],
+        ]
+
+        count = db.archive_and_compact("sess1", SUMMARY)
+        live = db.get_messages("sess1")
+        assert [row["id"] for row in live] == first_ids
+        assert count == 2
+        compacted = db.get_messages("sess1", include_compacted=True)
+        summaries = [
+            row
+            for row in compacted
+            if row.get("content") == SUMMARY[0]["content"]
+        ]
+        assert len(summaries) == 1
+
     def test_concurrent_tail_survives_compaction(self, db: SessionDB) -> None:
         _seed(db)
         watermark = db.get_active_message_watermark("sess1")


Enough1122:

**Overall:** Fixes #97321 — when the live transcript already equals the compacted set (e.g., a second compression path in the same cycle: turn preflight + tool-loop tail), `archive_and_compact` now short-circuits instead of archiving then re-inserting an identical row as compacted=1, which produced a duplicate visible via `session_search`. Clean, well-scoped guard with a regression test.

**What it does**
- Before archive/insert, `_do(conn)` calls `_active_messages_match_compacted(conn, session_id, compacted_messages)`: compares live active rows (role + content, in `id` order) against the compacted list; on a match it returns the current active-message count and skips the archive/insert entirely.
- New `_active_messages_match_compacted` helper compares `role` and encoded `content` per row with `zip`, requiring equal length first.
- Test: a second identical `archive_and_compact` keeps the same message ids, returns the correct count, and leaves exactly one summary row (including compacted rows).

**Non-blocking notes**
- The comparison requires exact match including order (`ORDER BY id` + `zip`) and content encoding via `_encode_content`. That's the conservative direction (a near-match falls through to the normal archive path), which is safe. The only risk is the inverse: if `_encode_content` ever diverges from how rows are actually stored (e.g., dict serialization differences), the guard silently stops triggering — the test locks the round-trip, so this is low risk.
- The early-return path returns `int(count_row["n"])` to preserve the function's active-count contract; the test asserts `count == 2` for a two-message summary, so the contract is pinned. If the function's return value semantics ever change elsewhere, this branch must be updated in lockstep.
- Consider `index=False`/cursor-named-field robustness — the code already guards `count_row is not None`; the `row["role"]`/`row["content"]` access assumes the connection returns dict-like rows, which matches how the surrounding code reads them.

PR #97627, opened 2026-08-29
fix(guard): evaluate trailing consecutive attempts in deterministic empty detection

ericmaddox:
## What does this PR do?

Fixes an empty response retry loop and cost-runaway bug in [`agent/empty_response_guard.py`](file:///c:/Users/EricM/Dev/hermes-agent/agent/empty_response_guard.py) where `deterministic_empty(agent)` evaluated all historical streak attempts (`all(a.usage_present ... for a in attempts)`) rather than inspecting the trailing consecutive attempts (`attempts[-2:]`).

In Hermes Agent:
1. When a model returns zero output tokens deterministically (e.g. portal-proxied silent refusals or prompt-triggered empty responses), `deterministic_empty(agent)` is designed to detect two consecutive zero-output attempts, skip further costly retries on that model, and immediately walk the fallback model chain.
2. Previously, `deterministic_empty(agent)` checked `all(a.usage_present and a.zero_output and a.signature == first.signature for a in attempts)`.
3. If a streak started with a transient attempt (such as missing usage headers or a single decoding token on attempt 0), and was followed by two consecutive deterministic zero-output completions on attempts 1 and 2, `all(...)` evaluated to `False` because `attempts[0]` failed the condition.
4. The transient first attempt permanently poisoned the entire streak, disabling deterministic empty detection. The agent was forced to execute all retries (with up to 60-second backoffs), re-sending full conversation context repeatedly at high cost while delaying fallback model activation.

The fix updates `deterministic_empty(agent)` to inspect the trailing two consecutive attempts (`attempts[-2:]`), ensuring that consecutive deterministic empties trigger immediately while transient attempts still fail open.

## Related Issue

N/A

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `agent/empty_response_guard.py`: Updated `deterministic_empty()` to evaluate `last_two = attempts[-2:]` for usage presence, zero output, and matching `(model, provider, finish_reason)` signatures.
- `tests/agent/test_empty_response_guard.py`: Added `test_nonzero_then_two_zero_attempts_is_deterministic` and `test_missing_usage_then_two_zero_attempts_is_deterministic` asserting that transient initial attempts do not poison subsequent consecutive zero-output detections.

## How to Test

Run the empty response guard test suite:
```bash
uv run --with pytest pytest tests/agent/test_empty_response_guard.py -v
```

### Code

- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(guard): evaluate trailing consecutive attempts in deterministic empty detection`)
- [x] I've tested on my platform: Windows 11 / Linux

## Screenshots / Logs

```text
============================= test session starts =============================
platform win32 -- Python 3.11.15, pytest-9.1.1, pluggy-1.6.0
rootdir: hermes-agent
configfile: pyproject.toml
plugins: anyio-4.12.1
collected 35 items

tests/agent/test_empty_response_guard.py ................................... [100%]

============================= 35 passed in 4.58s ==============================
```

### DIFF
--- a/agent/empty_response_guard.py
+++ b/agent/empty_response_guard.py
@@ -240,21 +240,22 @@ def record_empty_attempt(
 def deterministic_empty(agent: Any) -> bool:
     """True when the current streak looks deterministic.
 
-    Requires >= 2 consecutive attempts with an identical (model, provider,
-    finish_reason) signature. Usage-backed attempts must all prove zero output.
-    Usage-absent attempts must all have no observed content or reasoning. Mixed
-    evidence fails open so ambiguous transients keep their retries.
+    Requires >= 2 consecutive attempts, with the trailing attempts having an
+    identical (model, provider, finish_reason) signature. Usage-backed attempts
+    must all prove zero output. Usage-absent attempts must all have no observed
+    content or reasoning. Mixed evidence fails open so ambiguous transients
+    keep their retries.
     """
     if not guard_enabled(agent):
         return False
     attempts = getattr(agent, _ATTEMPTS_ATTR, None) or []
     if len(attempts) < 2:
         return False
-    first = attempts[0]
-    same_signature = all(a.signature == first.signature for a in attempts)
-    usage_proves_empty = all(a.usage_present and a.zero_output for a in attempts)
+    last_two = attempts[-2:]
+    same_signature = last_two[0].signature == last_two[1].signature
+    usage_proves_empty = all(a.usage_present and a.zero_output for a in last_two)
     response_proves_empty = all(
-        not a.usage_present and not a.observed_generation for a in attempts
+        not a.usage_present and not a.observed_generation for a in last_two
     )
     return same_signature and (usage_proves_empty or response_proves_empty)
 
--- a/tests/agent/test_empty_response_guard.py
+++ b/tests/agent/test_empty_response_guard.py
@@ -113,6 +113,32 @@ class TestDeterministicEmpty:
         )
         assert guard.deterministic_empty(agent) is False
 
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Fixes an empty-response retry/cost-runaway bug where `deterministic_empty` evaluated the *entire* historical attempt streak instead of the trailing consecutive attempts. A single transient early attempt (nonzero output or missing usage) previously poisoned the whole streak and blocked deterministic detection, causing repeated retries. Now only the last two attempts matter. Small, correct fix with two good regression tests.

**What it does**
- `agent/empty_response_guard.py:deterministic_empty` now takes `attempts[-2:]`, requires both to have usage present + zero output, and compares their `(model, provider, finish_reason)` signatures to each other (rather than to `attempts[0]`).
- Docstring updated to describe the trailing-consecutive semantics.
- Tests: `nonzero-then-two-zero-attempts` and `missing-usage-then-two-zero-attempts` both now report deterministic.

**Non-blocking notes**
- The signature comparison is between the last two attempts only; earlier streak members with a *different* signature no longer block detection. That's the intended "transient doesn't poison" behavior, but it means a signature flip then two silent empties will trip the guard even if the current signature is new — acceptable, since two consecutive zero-output attempts is a strong signal regardless.
- When `len(attempts) == 2`, `attempts[-2:]` is the whole list, so behavior is identical to before for the shortest streak — the change only affects longer streaks. Nice property, worth noting in the test suite maybe.
- The `len(attempts) < 2` early return is unchanged, so the fail-open minimum of 2 consecutive attempts is preserved.

PR #97632, opened 2026-08-29
fix(gateway): preserve provider context length on hot reload

Genion:
## Summary

- preserve provider/model-scoped `context_length` during Desktop/TUI live compression reloads
- include provider/model context pins in the live compression config signature so edits apply on the next turn
- keep top-level `model.context_length` precedence only when the active model and normalized route match
- support named custom providers, legacy `custom_providers`, omitted `model.base_url`, and dict-valued `model.default`
- invalidate all context-window-derived caches, including the summary token budget, when a pin is removed

## Problem

Agent initialization already resolves `providers.<provider>.models.<model>.context_length`, but the Desktop/TUI hot-reload path only considered top-level `model.context_length`.

With a provider-scoped pin such as:

```yaml
providers:
  example:
    models:
      gpt-5.6-sol:
        context_length: 256000
```

initialization correctly set the compressor to `256000`. On the next Desktop/TUI live sync, the provider pin was treated as absent and cleared. If endpoint metadata did not expose a context window, the compressor then fell back to the hardcoded model value (for the reproduced model, `1050000`). This affected both the displayed capacity and the actual compression threshold.

The fix makes the live path follow the same model/route scoping rules as initialization and covers sibling cases where a session is pinned to a different provider route.

## Verification

- `python -m pytest -q tests/tui_gateway/test_compression_config_hot_reload.py` — **19 passed**
- `python -m ruff check tui_gateway/server.py tests/tui_gateway/test_compression_config_hot_reload.py` — passed
- `python -m py_compile tui_gateway/server.py tests/tui_gateway/test_compression_config_hot_reload.py` — passed
- `python scripts/check-windows-footguns.py tui_gateway/server.py tests/tui_gateway/test_compression_config_hot_reload.py` — passed
- `git diff --check` — passed
- no-inference runtime probe with a provider-scoped `256000` pin — `before=256000`, `after=256000`

I also ran the broader `tests/tui_gateway` directory on Windows. Five unrelated failures appeared outside the changed files; four reproduce directly on a clean `origin/main`, and the remaining toolset assertion passes in isolation and is order-dependent. The focused regression file and all static gates above are clean.

## Platform

- Windows 11
- Python 3.11.16

### DIFF
--- a/tests/tui_gateway/test_compression_config_hot_reload.py
+++ b/tests/tui_gateway/test_compression_config_hot_reload.py
@@ -82,6 +82,31 @@ def test_unchanged_compression_config_is_noop(monkeypatch):
     assert compressor.threshold_tokens == 99_999
 
 
+def test_provider_model_context_change_invalidates_live_compression_signature():
+    def _cfg(context_length):
+        return {
+            "model": {
+                "default": "gpt-5.6-sol",
+                "provider": "relay",
+                "base_url": "https://relay.example.test/v1",
+            },
+            "providers": {
+                "relay": {
+                    "api": "https://relay.example.test/v1",
+                    "models": {
+                        "gpt-5.6-sol": {"context_length": context_length},
+                    },
+                },
+            },
+            "compression": {},
+        }
+
+    before = server._tui_compression_config_signature(_cfg(256_000))
+    after = server._tui_compression_config_signature(_cfg(192_000))
+
+    assert before != after
+
+
 def test_clearing_threshold_tokens_restores_ratio_trigger(monkeypatch):
     session, compressor = _session_with_compressor(threshold_tokens_cap=100_000)
     assert compressor.threshold_tokens == 100_000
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Fixes a Desktop/TUI hot-reload bug where a provider/model-scoped `context_length` pin (named custom providers, `providers.<name>.models.<model>.context_length`) was erased on the first live compression sync when top-level `model.context_length` was absent — and a route-mismatched global pin could inflate another model's compressor window. The change resolves the configured model/provider/base_url route, keeps global-pin precedence only when the active route matches, falls back to the per-model provider pin, and folds the relevant context metadata into the live-config signature. Substantial and well-tested.

**What it does**
- `_tui_compression_config_signature` now includes non-secret `providers.<name>.models.<model>.context_length` values (identity + base_url + model + repr(ctx)), so editing the active model's pin is picked up on the next turn.
- `_apply_live_compression_config`: handles dict-valued `model.default` via `split_model_config_default`; resolves `configured_base_url` from `model.base_url` or by matching custom-provider runtime ids; consults `should_clear_context_pin` so a top-level `model.context_length` only applies when the active model/route matches (fail-closed on unprovable identity); falls back to `get_custom_provider_context_length` when no global pin applies; and now also sets `agent._config_context_length` plus resets `cc._max_summary_tokens`/`_resolved_context_length` on clear.
- 5 new tests: signature invalidation on provider/model context change; removing context_length re-infers summary budget; provider pin survives live sync without a global pin; global pin doesn't override a different active custom route; global pin wins on a same-named route without a model URL.

**Non-blocking notes**
- `repr(metadata.get("context_length"))` in the signature means the same numeric value written as `256000` vs `"256000"` (int vs string) yields a different signature → spurious invalidation. Harmless (idempotent re-sync), but `int(...)`-normalizing in the signature would make it representation-stable.
- The signature includes context pins for *all* providers/models, not just the active one — any unrelated provider's pin edit also invalidates live compression config. Safe but broader than necessary; if that ever churns, scoping to the active route would tighten it.
- Broad `except Exception` around the route-matching and provider lookups fail closed toward "no pin", which is the safe direction — but it also silently swallows genuine config-parsing bugs. A debug log in those branches would help future diagnosis.
- New `agent._config_context_length` attribute is set on whatever object `agent` is (duck-typed in tests via `SimpleNamespace`); confirm the real agent class tolerates setting an attribute it may not declare (Python allows it, but it's worth being intentional about).

PR #82078, opened 2026-08-08
fix(desktop): archive kanban boards from Board settings

Per0-1:
## Summary
Desktop Kanban **Board settings** only allowed renaming and project binding. There was no way to remove a board from the Desktop UI, even though:

- browser dashboard already has **Archive** for non-`default` boards
- backend already exposes `DELETE /api/plugins/kanban/boards/{slug}` (archive by default)
- CLI already has `hermes kanban boards rm <slug>`

This adds Desktop parity: **Archive board** in Board settings (destructive, confirmed), hidden on `default`. Hard-delete remains CLI-only (`--delete` / `?delete=true`).

## Changes
- `apps/desktop/src/plugins/kanban/api.ts` — `archiveBoard(slug)` → `DELETE /boards/{slug}`
- `apps/desktop/src/plugins/kanban/board-switcher.tsx` — Archive button + `ConfirmDialog`; switch away if the active board was archived
- `apps/desktop/src/plugins/kanban/i18n.ts` — en/ja/zh/zh-hant strings

## Test plan
- [ ] Open Kanban → board switcher → **Board settings…** on a non-default board
- [ ] Confirm **Archive board** appears (left, destructive) and is absent on `default`
- [ ] Cancel the confirm dialog → board unchanged
- [ ] Confirm archive → board leaves the switcher list; if it was active, UI falls back to the server current board (usually `default`)
- [ ] Board dir lands under `~/.hermes/kanban/boards/_archived/`
- [ ] `hermes kanban boards list` no longer shows the archived slug

PR #97636, opened 2026-08-29
Kanban board export / import, plus board rename and delete in the desktop switcher

OutThisLife:
## Summary

A kanban board could be created and configured but never moved, and the desktop switcher had no way to rename or remove one. This adds whole-board transfer — `hermes kanban boards export|import`, matching REST endpoints, and Export/Import in the switcher menu — and fills in the missing rename and delete.

A board is not a row you can dump; it is a directory tree around a live SQLite database. Two things follow. **The database is live:** kanban runs in WAL mode, so a filesystem copy loses whatever still sits in the `-wal` sidecar and tears outright if the dispatcher commits mid-copy. Export goes through SQLite's online-backup API instead. **Rows carry machine-local state:** claims, worker PIDs, absolute workspace and attachment paths, session ids, and the gateway chat ids subscribed to task events. Shipping those verbatim is how an imported board arrives holding a claim owned by a process on someone else's laptop, or starts pushing task events into a stranger's Telegram thread. Everything machine-local is stripped on export and stripped again on import, because an archive is untrusted input.

Imports always land as a **new** board, auto-suffixing the slug on collision, so an import can never merge into or overwrite a board you already have. Tasks whose workspace was a directory or git worktree on the source machine are parked in `triage` rather than left for the dispatcher to claim, fail to build, and burn into the failure breaker.

Delete archives rather than erases — the board's directory moves to `boards/_archived/` and the toast names the path — and is never offered for `default`, which the backend refuses to remove. Rename technically already existed, buried as a field inside "Settings…", which is why it read as missing; it now has its own entry and the settings dialog is left owning scope alone.

Two pieces of supporting work came out of this. Profile export/import owned the only hardened tar handling in the tree, so its four helpers move to `hermes_cli/archive_safe` rather than getting a second, subtly weaker copy. And `PluginOs` had no way to ask for a file path, so plugins needing a dialog had to reach around the SDK for `window.hermesDesktop`; it now carries `pickSavePath` / `pickOpenPath`.

The last commit is a separate concern that surfaced while labelling the new menu: per-item menus across the app say "Rename", "Delete", "Export", but a handful of places had drifted to verb+noun or Title Case, so the same action read differently depending on where you found it. Nouns are kept where they carry weight — dialog titles, icon-button tooltips, `Remove worktree` (its menu also has a plain `Remove`), and `Open Bot Chat`, which names the canonical session titled exactly that.

## Test plan

- [ ] `scripts/run_tests.sh tests/hermes_cli/test_kanban_transfer.py tests/plugins/test_kanban_board_lifecycle_api.py tests/hermes_cli/test_profiles.py` — 79 passed locally
- [ ] `cd apps/desktop && npx vitest run src/plugins/kanban src/contrib/plugin.test.ts src/i18n`
- [ ] `cd apps/desktop && npx tsc --noEmit && npx eslint src`
- [ ] Round-trip a board with attachments between two `HERMES_HOME`s; confirm the import lands as a new slug and carries tasks, comments, and files
- [ ] Import onto a home that already has that slug; confirm the suffix and the "imported as" toast
- [ ] Export a board mid-dispatch; confirm the archive has no claims, PIDs, or `kanban_notify_subs` rows
- [ ] Rename a board from the switcher; confirm the slug is unchanged and the trigger label updates
- [ ] Delete a non-default board; confirm it archives to a restorable directory and the switcher falls back to `default`
- [ ] Confirm `default` never shows the delete entry


alt-glitch:

Related: #82078 overlaps on non-default board archival in the desktop UI; this PR additionally introduces safe board export/import and dedicated rename/delete actions. Please consolidate the overlapping UX/API decisions during review.

Issue #85700, opened 2026-08-13
Feature Request: Add enabled_toolsets parameter to delegate_task for per-subagent tool restriction


AIstudent04:
## Problem

`delegate_task` subagents inherit the parent's full toolset unconditionally. When a parent spawns a focused worker (e.g. a research subagent that only needs `web` and `file`, or a reviewer that only needs `file` and `vision`), every token-consuming tool description is still injected into the child's system prompt — `browser`, `image_gen`, `computer_use`, `tts`, `spotify`, `homeassistant`, etc. — none of which the child will ever call.

This wastes context on every turn of every subagent, adding up quickly in fleet/kanban workflows where subagents are spawned continuously.

## Requested Solution

Add an `enabled_toolsets` parameter (and/or a `disabled_toolsets` parameter) to `delegate_task`, parallel to the existing `enabled_toolsets` field on cron jobs:

```python
delegate_task(
    goal="Research competitors",
    context="...",
    enabled_toolsets=["web", "file", "skills", "memory"]
)
```

When set, the child agent should receive ONLY the listed toolsets — no others. When omitted (None), behavior stays as today (full inheritance).

## Why now

With the kanban fleet system gaining traction and users running multi-profile fleets (builder, designer, ops, reviewer, research), per-agent tool scoping is the single biggest token-saving lever available. Most specialist profiles only use 4–5 toolsets out of 20+.

## Existing precedence

- Cron jobs already have `enabled_toolsets` (though it has bugs — #82912, #53416, #80803)
- Kanban dispatched workers already gate to a focused `kanban_*` toolset via `HERMES_KANBAN_TASK`
- Per-profile `agent.disabled_toolsets` exists for the top-level session
- Per-invocation `-t / --toolsets` CLI flag exists for spawned Hermes processes

The missing piece is plumbing this through to `delegate_task` children.

## Alternatives considered

- Using `agent.disabled_toolsets` in profile config.yaml: works for top-level sessions but `delegate_task` children aren't full session starts — they share the parent's tool registry
- Spawning full Hermes processes with `-t` flag: works but is heavier than delegation for quick subtasks

PR #83019, opened 2026-08-10
fix(plugins): prevent duplicate correlated subagent launches

fangliquanflq:
## What does this PR do?

Concurrent plugin callbacks can currently launch the same correlated subagent job twice, duplicating delegated work and side effects while leaving only one child reachable through the correlation map. This change atomically reserves each non-empty parent-scoped correlation ID before child construction and precisely rolls the reservation back when launch setup fails.

### Symptom

Two synchronized `SubagentLifecycleService.launch()` calls with the same parent and non-empty `correlation_id` can both return valid handles and execute distinct children. The shared correlation map retains only the later child ID.

### Impact

Plugins can run delegated work and its side effects twice even though the public lifecycle contract treats correlation IDs as unique per parent. Correlation-based status tracking then loses the association to one of the running children. The unmeasured production frequency depends on plugins issuing concurrent duplicate launches.

### Bug Cause

**Trigger:** `agent/subagent_lifecycle.py` / `SubagentLifecycleService.launch()` when concurrent calls share one parent-scoped non-empty correlation ID.

**Causal chain:**
1. One launch checks the correlation map and releases the registry lock before building its child.
2. A concurrent launch checks during that gap, also sees no mapping, and builds another child.
3. Both children are submitted, and the later correlation write overwrites the earlier association while both execute.

**Why it is wrong:** The duplicate check and claim were separated by unlocked child construction, so checking uniqueness did not reserve the key.

**Working sibling / contrast:** Sequential duplicates are already rejected after the first registration. Empty correlation IDs remain intentionally unrestricted, and equal values under distinct parents remain isolated by the parent-scoped key.

**Ruled out:** This is not caused by terminal cleanup or retention. The race occurs before either child reaches a terminal state and was reproduced with synchronized construction on the default-branch tip.

### Fix

- Reserve non-empty parent-scoped correlation IDs inside the initial registry critical section while keeping child construction outside the lock.
- Replace the reservation with the assigned child ID when construction succeeds.
- Remove only the exact reservation or record on import, construction, identity, or executor-submission failure so retries and unrelated keys remain valid.
- Add deterministic concurrency and rollback tests for duplicate launch, build failure, missing identity, and rejected executor submission.

## Related Issue

Fixes #83018

## Type of Change

- [x] Bug fix

## Changes Made

- `agent/subagent_lifecycle.py` - reserve correlation keys before child construction and make launch rollback transactional.
[...]

PR #68499, opened 2026-07-21
fix(delegation): separate lifecycle from task outcome

ibaldr89:
## Summary

- separate delegated child lifecycle status from logical task outcome
- propagate fail-closed outcome/evidence through sync and durable async completion paths
- render unverified/partial/failed/unknown states without false-green success signals in gateway, TUI, and Desktop
- classify current fatal/partial runtime exit reasons consistently, including budget/max-iteration, partial-stream, and abandoned-owner recovery cases

## Safety contract

- child self-report output is never treated as verified task acceptance
- authoritative runtime errors and fatal exits override optimistic outcome data
- abandoned recovery preserves `unknown`, including durable `exit_reason`, to avoid unsafe duplicate side effects
- legacy durable records fail closed while retaining usable partial output

## Validation

Exact candidate: `301a359ccbd922f616acb2091ff451aaf108fd8b`  
Tree: `8659ed8a79e4bfecaf996bfc4a25092da30974ea`  
Base/direct parent: `8cab422ab09332c1867af81ee4e910878ac1172b`

- exactly one surviving commit directly on pinned live `main`
- current main's UTF-8 sandbox-log poller fix (`75bcd8668`) is composed into bounded `tools/process_registry_runtime.py`; its exhaustive split-sequence regression remains in the base test suite
- stable patch ID: `dce1aa53c57f17fd04b08f7de3c4b0010c9d3fdd`
- all **46** modified files are <=2,000 physical lines (largest: `tools/delegate_tool.py`, **1,993**; hooks documentation: **1,934**)
- exact-tree Python process/delegation/lifecycle/schema/gateway slice: **245 passed, 5 skipped, 7 subtests passed**
- exact-tree Desktop outcome slice: **44 passed**; TUI outcome slice: **41 passed**
- Desktop and TUI typechecks, TUI Ink build, targeted ESLint, Ruff, compile checks, `git diff --check`, one-commit/direct-parent topology, contributor attribution, and redacted added-diff scan: pass
- exact remote guard passed immediately before publication: upstream `main` remained `8cab422a`, old branch head remained `45f4590e`, and post-push readback matched `301a359c`
[...]

PR #63359, opened 2026-07-12
feat(plugins): add public subagent lifecycle API

asimons81:
## Summary

Expose a typed, plugin-safe subagent lifecycle service through `PluginContext.subagent_lifecycle`. Plugins can launch, monitor, cancel, and retrieve results from fresh Hermes child sessions without importing private delegation internals, gateway state, or TUI components.

- Add `agent/subagent_lifecycle.py` -- `SubagentLifecycleService` with launch, status, bounded wait, cooperative cancellation, immutable terminal-result retrieval, and reconnect diagnostics
- Add `PluginContext.subagent_lifecycle` property (`hermes_cli/plugins.py`, +17 lines) with lazy initialization and a parent-agent resolver lambda
- Add `tests/agent/test_subagent_lifecycle.py` -- 98 lines of contract/security tests
- Add `website/docs/developer-guide/subagent-lifecycle-api.md` -- public API documentation

## Public API

The service is obtained from `PluginContext`:

```python
service = ctx.subagent_lifecycle
handle = service.launch(SubagentLaunchRequest(goal="..."))
status = service.status(handle)
terminal = service.wait(handle, timeout_seconds=30)
result = service.result(handle)
cancel_result = service.cancel(handle, reason="...")
reconnect_result = service.reconnect(handle)
```

`SubagentHandle` is fully serializable (`to_dict()` / `from_dict()`) and carries a versioned, opaque HMAC capability. Malformed or forged handles return `UNKNOWN` / `UNKNOWN_HANDLE` and cannot access a child.

Key request fields:

| Field | Type | Limit |
|---|---|---|
| `goal` | `str` | 16 000 chars |
| `context` | `Optional[str]` | 32 000 chars |
| `metadata` | `Mapping[str, Any]` | 8192 bytes JSON |
| `allowed_toolsets` | `Optional[tuple[str, ...]]` | Must be subset of parent toolsets |
| `correlation_id` | `Optional[str]` | Unique per parent session |

## Lifecycle semantics

States: `PENDING` -> `STARTING` -> `RUNNING` -> `SUCCEEDED` / `FAILED` / `INTERRUPTED`, with `CANCEL_REQUESTED` as a transient during cooperative cancellation. `UNKNOWN` is returned for unrecognized or forged handles.

- `cancel()` is cooperative: it calls `agent.interrupt()` and returns `CANCEL_REQUESTED`. The state only transitions to `CANCELLED` after terminal confirmation from the child loop.
[...]

PR #97637, opened 2026-08-29
fix(lifecycle): enforce parent disabled_toolsets in subagent launch validation

ericmaddox:
## What does this PR do?

Hardens `SubagentLifecycleService.launch()` validation (`agent/subagent_lifecycle.py`) to reject requested `allowed_toolsets` that intersect the parent agent's `disabled_toolsets` before child construction (fail-fast contract parity at the public service boundary).

While `model_tools._compute_tool_definitions()` already subtracts inherited disabled toolsets downstream during child assembly, enforcing this at `SubagentLifecycleService._validate_request()` provides deterministic, fail-fast rejection of requests that attempt to broaden parent permissions.

## Related Issue

N/A (public lifecycle service boundary hardening)

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `agent/subagent_lifecycle.py`:
  - In `_validate_request()`: check if `request.allowed_toolsets` intersects `parent.disabled_toolsets` and raise `SubagentLifecycleError` with clear diagnostic text before child construction.
- `tests/agent/test_subagent_lifecycle.py`:
  - Added `test_launch_rejects_toolsets_in_parent_disabled_toolsets` asserting that:
    1. A forbidden toolset request raises `SubagentLifecycleError` and the child builder is never invoked.
    2. A disjoint allowed request successfully reaches the child builder with the expected toolset and parent reference.

## How to Test

1. Run lifecycle test suite:
   ```bash
   pytest tests/agent/test_subagent_lifecycle.py -v
   ```

### Code

### DIFF
--- a/agent/subagent_lifecycle.py
+++ b/agent/subagent_lifecycle.py
@@ -414,3 +414,6 @@ class SubagentLifecycleService:
         enabled = getattr(parent, "enabled_toolsets", None)
         if enabled is not None and not set(request.allowed_toolsets).issubset(set(enabled)):
             raise SubagentLifecycleError("Requested toolsets would broaden parent permissions.")
+        disabled = getattr(parent, "disabled_toolsets", None)
+        if disabled and (set(request.allowed_toolsets) & set(disabled)):
+            raise SubagentLifecycleError("Requested toolsets would broaden parent permissions.")
--- a/tests/agent/test_subagent_lifecycle.py
+++ b/tests/agent/test_subagent_lifecycle.py
@@ -178,3 +178,52 @@ def test_agent_turn_binds_and_clears_lifecycle_parent(monkeypatch):
     assert agent.run_conversation("hello") == {"final_response": "ok"}
     assert observed == [agent]
     assert get_active_subagent_parent() is None
+
+
+def test_launch_rejects_toolsets_in_parent_disabled_toolsets(monkeypatch):
+    """Requesting toolsets disabled on the parent raises SubagentLifecycleError before child construction."""
+    parent = SimpleNamespace(
+        session_id="parent-sec",
+        enabled_toolsets=None,
+        disabled_toolsets=["terminal", "web"],
+    )
+    service = SubagentLifecycleService(lambda: parent)
+
+    captured_builds = []
+
+    def mock_build_child(**kwargs):
+        captured_builds.append(kwargs)
+        return FakeChild("sa-sec-test")
+
+    monkeypatch.setattr(
+        "tools.delegate_tool._build_child_preserving_parent_tools",
+        mock_build_child,
+    )
+    monkeypatch.setattr(
+        "tools.delegate_tool._run_single_child",
+        lambda *_args, **_kwargs: {"status": "completed"},
+    )
+
+    # Denied: toolset in parent's disabled_toolsets -> fails fast before child builder
+    with pytest.raises(
+        SubagentLifecycleError, match="Requested toolsets would broaden parent permissions"
+    ):
+        service.launch(
+            SubagentLaunchRequest(
+                goal="do something",
+                allowed_toolsets=("terminal", "file"),
+            )
+        )
+    assert len(captured_builds) == 0, "Child builder must not be invoked when validation fails"
+
+    # Allowed: toolset disjoint from parent's disabled_toolsets -> reaches child builder
+    handle = service.launch(
+        SubagentLaunchRequest(
+            goal="do something safe",
+            allowed_toolsets=("file",),
+        )
+    )
+    assert len(captured_builds) == 1
+    assert captured_builds[0]["toolsets"] == ["file"]
+    assert captured_builds[0]["parent_agent"] is parent
+    assert handle.subagent_id == "sa-sec-test"


andrexibiza (commented):
Reviewed exact head `6b36e207cc28b8381f068a53b1b9b15f3d9da7d4`. The PR's comparison base is `36620578f09f2a716102c845dc3b1368d6f761a7`; live `main` is now `ee742fe1bc828f6456659f67d27bdbeacdebbdd4` (three commits ahead of that base, with no `agent/subagent_lifecycle.py` delta in that interval). I also traced the public lifecycle path through `_build_child_preserving_parent_tools` / `_build_child_agent`, the tool resolver, the merged lifecycle origin (#72501, preserving #63359's contributor lineage), exact-head Actions, and the active delegation-outcome work in #68499.

The fail-fast validation itself is a reasonable hardening layer, but I don't think this head currently proves the security bug described in the PR. I would not merge it under the present privilege-escalation claim without tightening both the evidence and the test.

### Blocker — the claimed `terminal` / `web` privilege escalation is already denied at the actual child tool-resolution boundary on this PR's base

`SubagentLifecycleService.launch()` does pass the requested `allowed_toolsets` into `_build_child_preserving_parent_tools`, but it also passes the live `parent_agent`. On the exact base, `_build_child_agent()` already reads `parent_agent.disabled_toolsets`, carries those values into `child_disabled_toolsets`, and supplies them to `AIAgent` together with the requested enabled set.

More importantly, `model_tools._compute_tool_definitions()` explicitly applies disabled toolsets **as a final subtraction even when enabled toolsets are present**. The code comment ties that invariant to #17309. The existing base test `tests/tools/test_delegate.py::TestStripBlockedTools::test_mixed_composite_is_subtracted_at_child_assembly` exercises the real assembly shape and verifies the deny list survives composite expansion before the child receives tools.

So for the concrete example in the PR body — parent `disabled_toolsets=["terminal"]`, plugin request `allowed_toolsets=("terminal",)` — the request currently passes `_validate_request`, but the child does **not** thereby acquire terminal access. This patch changes *where* the request is rejected; it does not, from the repository evidence I can find, close the described sink-level privilege escalation.

Please do one of two things before this lands:

1. If there is a real escape despite the inherited deny path, add a regression that is RED on `36620578...` and demonstrates the forbidden tool is actually present/dispatchable on the child, then GREEN on this head; or
2. Reframe this accurately as public-boundary fail-fast / contract-parity hardening: reject an impossible broadening request before child construction even though the child resolver would later subtract the denied tools. In that case the PR body/security classification should stop claiming an access escalation that the target base already prevents.

That distinction matters here because this API's original contract (#72501, salvaging #63359 with attribution preserved) explicitly says `allowed_toolsets` narrows and never widens. We should keep the public boundary honest, but not turn earlier validation into evidence of a vulnerability at a later sink.

### Blocker — the positive control in the new test accepts any downstream failure

The second half of `test_launch_rejects_toolsets_in_parent_disabled_toolsets` does:

```python
with pytest.raises(Exception) as exc_info:
    service.launch(... allowed_toolsets=("file",))
```

and then only asserts that the arbitrary exception is not this particular broadening error. That means import failure, child-construction failure, executor failure, unrelated lifecycle regression, etc. all make the "allowed" case pass. It does not prove a disjoint request crosses validation correctly.

Please make this deterministic: stub/capture the child-build boundary (or complete a controlled launch) and assert the allowed request actually reaches the expected post-validation operation. For the denied case, also assert the child builder is never called. That gives us a precise pre-effect witness rather than using an unrelated exception as the oracle.

### Other side of the shape: this is not a general `disabled_toolsets` monotonicity proof

The shared delegate builder has an intentional orchestrator exception: when depth/config grants orchestrator capability, it removes `"delegation"` from inherited disabled toolsets and re-adds the delegation toolset. `test_orchestrator_composite_regains_only_delegate_task` codifies that policy. I am **not** asking this PR to redesign that behavior, but the scope needs to be stated precisely: this patch rejects explicit `allowed_toolsets ∩ parent.disabled_toolsets`; it does not establish that every parent disabled-toolset decision is monotonic across every child construction path.

### Interlocks / merge order / acceptance

- #72501 is the merged salvage/integration of #63359 and preserves the original contributor's lifecycle work; this PR is adjacent hardening of that API, not a replacement for that lineage.
- #68499 is active and modifies both `agent/subagent_lifecycle.py` and `tests/agent/test_subagent_lifecycle.py` while tightening lifecycle/outcome evidence. It is complementary, not duplicate work, but whichever lands second needs a fresh rebase plus the combined lifecycle suite because these are direct file collisions on a security-sensitive boundary.
- #83019 is a closed, unmerged historical correlation-reservation repair on the same lifecycle surface; useful precedent for deterministic launch-boundary tests, but not an active competing owner.
- #85700 is an open per-subagent tool-restriction feature request; adjacent policy surface, not a duplicate of this validation change.

Exact-head acceptance is also not green yet. CI `33231914083`, Docker `33231913651`, and Nix `33231913654` are all `action_required`, and each currently has zero jobs. The contributor's local 5/5 focused result is useful, but it is not an executable exact-head repository receipt.

The narrow guard is sensible and worth keeping if it is described and proved for what it actually does. The main ask is to make the security story exact: prove the sink escape if one exists, otherwise ship this as fail-fast defense in depth with a deterministic boundary test. Nice catch on the missing early validation path; getting the proof aligned with the real enforcement stack will make this much stronger.

ericmaddox:
Thanks for the thorough review and tracing through `_compute_tool_definitions` and the delegate builder, @andrexibiza!

You're completely right on the distinction between the public API boundary contract and the downstream tool-resolution sink. I've updated the branch and PR accordingly:

1. **Reframed to Fail-Fast Contract Parity / Boundary Hardening**: Updated the PR framing to clearly position this as public boundary fail-fast validation adhering to `#72501`'s invariant (`allowed_toolsets` may only narrow, never broaden), rejecting impossible broadening requests before child construction rather than relying on downstream sink subtraction.
2. **Deterministic Pre-Effect Test (`4786c135b`)**: Refactored `test_launch_rejects_toolsets_in_parent_disabled_toolsets` in `tests/agent/test_subagent_lifecycle.py` to mock `_build_child_preserving_parent_tools`. It now deterministically verifies:
   - For forbidden toolsets: `SubagentLifecycleError("Requested toolsets would broaden parent permissions.")` is raised and the child builder is **never called** (`len(captured_builds) == 0`).
   - For allowed toolsets: the child builder **is called** with the expected arguments (`toolsets=["file"]`).

All unit tests pass cleanly locally (`5/5 passed in tests/agent/test_subagent_lifecycle.py`). Ready for your re-review!

ericmaddox:
@andrexibiza Thank you for the review feedback!

I have updated the PR and rebased onto current `main` (`a6e10e693f`):
1. **Accurate Reframing**: Updated the PR title and description to accurately reflect this as **public-boundary fail-fast contract hardening** ensuring `_validate_request()` rejects impossible broadening requests before child construction (matching the contract of downstream child tool resolution).
2. **Deterministic Boundary Test**: Refactored `test_launch_rejects_toolsets_in_parent_disabled_toolsets` in `tests/agent/test_subagent_lifecycle.py` with explicit spy capture:
   - For denied toolsets intersecting `parent.disabled_toolsets`: Asserts `SubagentLifecycleError` is raised and verifies the child builder is never invoked.
   - For allowed disjoint toolsets: Asserts the request cleanly reaches the child builder with the expected toolsets and parent reference (5/5 tests passing).

PR #72742, opened 2026-07-27
fix(discord): release the agent when clarify buttons expire

hyuseinleshov:
## What does this PR do?

A native multi-choice `clarify` on Discord is rendered by `ClarifyChoiceView`, whose lifetime comes from `approvals.discord_prompt_timeout` (**300 s** by default). The agent thread that asked the question is parked in `clarify_gateway.wait_for_response`, governed by a completely separate deadline — `agent.clarify_timeout` (**3600 s** by default, *unlimited* when set to `0`).

When the view timed out, `on_timeout` disabled the buttons and repainted the embed with **"⏱ Prompt expired — no action taken"** — and that was all. The clarify entry stayed armed, so the agent stayed blocked for the remaining ~55 minutes.

The prompt is genuinely unanswerable during that window:

- the buttons are disabled, so there is no click path left;
- `_coerce_text_response` rejects prose for an `awaiting_text=False` entry, so typing an answer does nothing (correct in itself — strict multi-choice coercion is deliberate);
- the rejected text falls through to the busy path, where `busy_input_mode: interrupt` calls `AIAgent.interrupt()`. That flag is only observed at loop checkpoints, and the thread is asleep inside `Event.wait`. So the message is queued behind the very turn it was meant to unblock.

Net effect: the session is pinned, the UI claims the prompt is dead, and every follow-up message vanishes silently — no `inbound message` log line is ever emitted for it.

### Evidence from a live gateway

Discord showed `⏳ Working — 54 min — iteration 7/150, clarify` under an embed reading "Prompt expired — no action taken". Three user messages (including *"Cancel the current task"*) were accepted by the adapter but never reached the agent:

```
13:32:42  agent.conversation_loop: API call #7 ... latency=11.9s   ← clarify raised here
13:48:14  [Discord] Flushing text batch ... (53 chars)             ← no `inbound message` line
14:14:52  [Discord] Flushing text batch ... (21 chars)             ← no `inbound message` line
14:20:06  [Discord] Flushing text batch ... (63 chars)             ← no `inbound message` line
```

`kill -USR2` on the gateway confirmed where the turn was stuck:

```
tools/clarify_gateway.py:139 in wait_for_response      ← Event.wait
gateway/run.py:21911 in _clarify_callback_sync
tools/clarify_tool.py:102 in clarify_tool
agent/tool_executor.py:1360 in _execute
```

## Related Issue

[...]

PR #97643, opened 2026-08-29
fix(discord): align clarify button lifetime

apellis:
## What does this PR do?

Discord native-choice `clarify` prompts currently have two independent lifetimes:

- `ClarifyChoiceView` uses `approvals.discord_prompt_timeout` (300 seconds by default, capped at 900 seconds).
- The gateway waiter uses `agent.clarify_timeout` (3600 seconds by default; `<= 0` means unlimited).

After five minutes, Discord disables the buttons and says the prompt expired even though Hermes may still be waiting for up to another 55 minutes—or forever when clarify is configured as unlimited.

This PR makes the Discord clarify view use the same canonical clarify timeout as the gateway waiter:

- positive timeout: the buttons and waiter share the same lifetime;
- non-positive timeout: `discord.ui.View(timeout=None)`, matching the documented unlimited wait;
- execution approvals and other security-sensitive Discord views continue using the separately bounded `approvals.discord_prompt_timeout`.

It also fixes the cancellation sentinel for an unlimited wait. A session interruption previously produced `[user did not respond within 0m]`; it now reports `[clarify wait cancelled before user response]`.

## Related work

Related but intentionally different from #72742.

#72742 keeps the five-minute Discord view deadline, then introduces a second typed-answer grace timer and eventually resolves the waiter with an empty response. This PR instead removes the deadline split at its source. That preserves the documented `agent.clarify_timeout <= 0` unlimited contract and avoids another timer/task/race lifecycle.

## Type of change

- [x] Bug fix

## Changes

- Add `_read_discord_clarify_timeout()` using the canonical clarify resolver.
- Use it only for `ClarifyChoiceView`; approval/confirmation views stay bounded.
- Distinguish unlimited-wait cancellation from a finite timeout.
- Add regressions for unlimited Discord view lifetime and the false `0m` sentinel.

## Test plan

The new tests were first run against the old behavior and failed as expected:

- Discord view timeout was `300`, not `None`.
- unlimited-wait cancellation returned `[user did not respond within 0m]`.

After the fix:

```bash
scripts/run_tests.sh tests/gateway/test_discord_clarify_buttons.py -q -k 'TestClarifyChoiceViewConstruction'
# 3 passed

scripts/run_tests.sh \
  tests/gateway/test_discord_prompt_timeout_config.py \
  tests/tools/test_clarify_gateway.py \
  tests/gateway/test_clarify_send_timeout_ambiguity.py -q
# 47 passed
```

Live verification on Discord/Linux:

- `agent.clarify_timeout: 0`
- resolved clarify view timeout: `None`
- resolved execution-approval view timeout: `300`
- gateway restarted onto the patched checkout successfully.

## Checklist

- [x] Read `CONTRIBUTING.md`
- [x] Searched open and closed issues/PRs
- [x] Cross-referenced the overlapping open PR instead of claiming no prior work
- [x] Branch contains only the four scoped code/test files
- [x] Added red-capable regression tests
- [x] Ran focused canonical test commands
- [x] Preserved bounded execution-approval behavior

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -986,6 +986,8 @@ def _clarify_send_then_wait(fut, *, clarify_id: str, session_key: str, clarify_m
     response = clarify_mod.wait_for_response(clarify_id, timeout=float(timeout))
     if response is None or response == "":
         # Timeout or session-boundary cancellation
+        if timeout <= 0:
+            return "[clarify wait cancelled before user response]"
         return f"[user did not respond within {int(timeout / 60)}m]"
     return response
 
--- a/plugins/platforms/discord/adapter.py
+++ b/plugins/platforms/discord/adapter.py
@@ -1031,6 +1031,24 @@ def _read_discord_prompt_timeout() -> int:
     return seconds
 
 
+def _read_discord_clarify_timeout() -> Optional[float]:
+    """Return the canonical clarify lifetime for Discord choice buttons.
+
+    Clarify prompts are not execution approvals: their button view must remain
+    usable for exactly as long as the gateway is waiting for the user's answer.
+    A non-positive canonical timeout means unlimited and maps to discord.py's
+    ``View(timeout=None)``. Security-sensitive approval views continue using
+    the separately bounded ``approvals.discord_prompt_timeout`` setting.
+    """
+    try:
+        from tools.clarify_gateway import get_clarify_timeout
+
+        seconds = float(get_clarify_timeout())
+    except Exception:
+        seconds = 3600.0
+    return None if seconds <= 0 else seconds
+
+
 class DiscordAdapter(BasePlatformAdapter):
     """
     Discord bot adapter.
@@ -9636,7 +9654,7 @@ def _define_discord_view_classes() -> None:
             allowed_user_ids: set,
             allowed_role_ids: Optional[set] = None,
         ):
-            super().__init__(timeout=_read_discord_prompt_timeout())
+            super().__init__(timeout=_read_discord_clarify_timeout())
             self.choices = list(choices)[:24]
             self.clarify_id = clarify_id
             self.allowed_user_ids = allowed_user_ids
--- a/tests/gateway/test_clarify_send_timeout_ambiguity.py
+++ b/tests/gateway/test_clarify_send_timeout_ambiguity.py
@@ -154,6 +154,21 @@ def test_no_response_returns_timeout_sentinel():
     )
 
 
+def test_unlimited_wait_cancellation_does_not_claim_zero_minute_timeout():
+    fut = MagicMock()
+    fut.result.return_value = _Result(True)
+    clarify_mod = MagicMock()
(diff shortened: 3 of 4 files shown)


Enough1122:

**Overall:** Aligns the Discord clarify button view's lifetime with the gateway's actual clarify wait. `ClarifyChoiceView` now uses `_read_discord_clarify_timeout()` (derived from the canonical `get_clarify_timeout()`, `None` when unlimited → `View(timeout=None)`) instead of the approval-specific `approvals.discord_prompt_timeout`, and the gateway's `_clarify_send_then_wait` returns a distinct cancellation sentinel for unlimited timeouts instead of a misleading "did not respond within 0m". Tests cover both sides.

**What it does**
- `_read_discord_clarify_timeout`: `None` when the canonical timeout ≤ 0 (unlimited), else the seconds; `3600.0` fallback on any exception. Security-sensitive approval views keep the separately bounded `approvals.discord_prompt_timeout`.
- `_clarify_send_then_wait`: when `timeout <= 0`, returns `"[clarify wait cancelled before user response]"` — avoids the nonsensical "within 0m" message on cancellation.
- Tests: unlimited clarify keeps Discord buttons live (`view.timeout is None`); unlimited-wait cancellation returns the new sentinel.

**Non-blocking notes**
- The bare `except Exception` in `_read_discord_clarify_timeout` silently falls back to 3600s. If `tools.clarify_gateway.get_clarify_timeout` is unavailable for a reason other than absence (import/config error), the view lifetime (3600s) can diverge from the configured `agent.clarify_timeout`. A `logger.debug` in the except branch would make such a divergence diagnosable.
- Behavior change worth confirming: with unlimited clarify, the Discord view now stays interactive with no expiry until the gateway resolves the prompt. That's the intended fix, but a stale card left on screen stays clickable indefinitely — verify the existing interaction/cleanup path still `stop()`s the view (and disables buttons) once the answer arrives.

sake303:
I added concrete lifecycle reproductions to the existing issue: https://github.com/NousResearch/hermes-agent/issues/97662#issuecomment-5568548482. I am consolidating my overlapping #104916 there rather than keeping another tracker open.

This also supports the earlier review's note about retiring resolved views. Against main source at `d8a07768c5ee59afae62e9fb51d45e1e30aa74da`, an offline test using the real discord.py ViewStore timer reproduced:

- An accepted clarify answer later overwritten with `Prompt expired — no action taken`.
- A missing clarify entry still rendered as `Answered` despite failed resolution.
- The completed-to-expired overwrite on execution-approval cards through the shared view lifecycle.

Would you prefer to include terminal-view cleanup and render-after-acceptance handling in this PR, or keep them as a coordinated follow-up under #97662? The issue comment includes pinned source anchors, test setup and limits, and proposed regression cases. Pending execution approvals should retain their separately bounded timeout.

This evidence is against the pinned main source, not a test of this PR's head. It supports the existing cleanup acceptance criteria without proposing a different timeout policy.

PR #97645, opened 2026-08-29
fix(honcho): omit reasoning-contaminated session summaries

### LINKED ISSUE #97639: [Bug]: Honcho Session Summary can persist model reasoning and stale cache reinjects it
## Bug Description

The bundled Honcho integration can persist and inject a generated **Session Summary that contains the summarizer model's own planning/reasoning text**, including a closing `</think>` marker.

The observed injected block began like this (private conversation details omitted):

```text
## Session Summary
I need to create a thorough, comprehensive summary of this conversation.
The instruction says to focus on capturing key facts...
First, let me review what I have in the previous summary...
...
</think>
<ordinary summary text>
```

This is derived memory, not user-authored content. Once stored, Hermes treats it as trusted background context and reinjects it into later turns. The failure can pollute answer planning, leak hidden model reasoning, and repeatedly surface obsolete or sensitive historical details.

This report is distinct from:

- #51511, which concerned unrelated error text appended during summary assembly;
- #93326, which concerns a phantom Desktop peer serving a stale but otherwise valid summary;
- #32858, which concerns background-curator instructions being misattributed as user preferences; and
- #10694, which concerns prompt-structuring delimiters from an untrusted memory provider.

### Production reproduction

1. Run Hermes Desktop with the self-hosted Honcho provider in `hybrid` or `context` recall mode.
2. Use a sufficiently long session for Honcho to generate short/long session summaries through a reasoning-capable summarizer model.
3. Allow a summarizer response containing planning prose and/or `</think>` to complete.
4. Inspect the session summaries returned by Honcho, or send another Desktop turn.
5. Observe that the malformed model output was persisted as a session summary and injected under `## Session Summary` inside `<memory-context>`.

In the affected workspace:

- 29 sessions were audited;
- 10 stored summary records were found;
- 7 contaminated summaries were present across 5 sessions;
- only 3 summaries were clean.

The contaminated records were surgically removed while raw messages, peer cards, representations, conclusions, memberships, and clean sibling summaries were preserved. After restarting only the Honcho API to clear its process cache, a public-API audit found 0 contaminated summaries and the Honcho queue was fully drained.

Despite that clean backend readback, a later turn in the already-running Desktop session again received the same stale malformed `## Session Summary` block. This is consistent with Hermes retaining the already-formatted block in `_base_context_cache` and reinjecting it until a refresh/process restart, without revalidating the cached content at the final injection boundary.

### Minimal consumer reproduction on unmodified v0.20.6

```python
provider = HonchoMemoryProvider()
formatted = provider._format_first_turn_context({
    "summary": (
        "I need to create a thorough, comprehensive summary.\n"
        "First, let me review what I have in the previous summary.\n"
        "</think>\n"
        "The user discussed their project."
    ),
    "representation": "The user prefers concise technical reports.",
    "card": "Role: developer",
})

assert "</think>" not in formatted  # fails on v0.20.6
```

The same malformed summary is also returned verbatim by the `honcho_context` tool path.

## Expected Behavior

- Model planning/reasoning text must never be persisted as a Honcho session summary.
- Hermes must treat all memory-provider summaries as untrusted derived data and reject/omit malformed summaries before injection.
- Rejection of a bad summary must preserve clean representation and peer-card context.
- Automatic prefetch and the explicit `honcho_context` tool must share the same summary-integrity check.
- Cached base context must not bypass the check. Once the backend summary is deleted or replaced, stale malformed content must not continue to be reinjected indefinitely.
- A rejected producer response should use a safe fallback or leave the previous clean summary unchanged; it must not overwrite memory with reasoning text.

## Actual Behavior

On upstream Hermes Agent v0.20.6:

1. `plugins/memory/honcho/__init__.py::_format_first_turn_context()` appends every non-empty summary verbatim:

```python
summary = ctx.get("summary", "")
if summary:
    parts.append(f"## Session Summary\n{summary}")
```

2. The explicit `honcho_context` path independently appends `ctx['summary']` verbatim.
3. The formatted string is stored in `_base_context_cache` and appended on later turns without a semantic-integrity check at the final injection boundary.
4. The Honcho summarizer accepted the reasoning-bearing model response as valid output and persisted it in `Session.internal_metadata["summaries"]`.

No exception or warning was emitted when the contaminated summary was persisted or injected.

## Affected Component

- Other: bundled Honcho memory plugin / self-hosted Honcho integration
- Desktop

N/A — reproduced in Hermes Desktop.

## Debug Report

Private, force-redacted Nous diagnostics bundle (auto-deletes after 14 days):

https://support.nousresearch.com/diagnostics/7afc9a4f-d3aa-4543-91aa-6a5c251f53a7

No public paste was created because the logs contain unrelated private conversation content and local file paths.

## Operating System

Windows 11 (`10.0.26200.9168`)

## Python Version

Python 3.11.15

## Hermes Version

Hermes Agent v0.20.6 (2026.8.27). Installed git HEAD: `ac6c8028e00d01ee2f299ba7fd03329c7f10382d`. The launcher reports release upstream `9a1eef7a` and **Up to date**.

There was no exception. The failure is invalid data being accepted and later trusted.

Sanitized audit result:

```text
sessions_scanned=29
summaries_before=10
contaminated_before=7
sessions_affected=5
clean_summaries_preserved=3
contaminated_after_api_restart=0
```

After repair/restart:

```text
honcho_health=ok
queue_total=44
queue_completed=44
queue_pending=0
queue_in_progress=0
```

A subsequent turn in the existing Desktop process still received the stale malformed summary block.

## Root Cause Analysis

There are three boundaries with no shared integrity invariant:

1. **Producer boundary:** the Honcho summary producer checked for empty output but did not reject reasoning-channel markers or known summary-planning preambles before writing derived state.
2. **Hermes consumer boundary:** `_format_first_turn_context()` and the `honcho_context` tool independently trust every non-empty summary string.
3. **Cache/injection boundary:** Hermes caches already-formatted base context in `_base_context_cache`. A contaminated cached string can be appended on later turns without being parsed or revalidated, including after the source summary has been repaired server-side.

The general `agent.memory_manager.sanitize_context()` fence sanitizer does not solve this semantic-integrity problem. It is designed around memory-context framing, while this failure is ordinary-looking prose plus a reasoning-channel marker inside a legitimate summary field.

## Proposed Fix

Use defense in depth at both ownership boundaries:

1. **Producer:** validate the generated summary before persistence. At minimum, reject `<think>` / `</think>` and known summary-planning output; preferably use a structured summary response contract and fail closed to a safe fallback when validation fails.
2. **Consumer:** add one provider-local `is_usable_session_summary()` / formatting helper and use it in both automatic prefetch and `honcho_context`. If the summary fails validation, omit only the summary while retaining clean representation/card context.
3. **Cache:** cache structured context or revalidate the formatted cache immediately before injection. A backend refresh that removes/replaces a summary must replace the old cache even when the newly formatted result is empty; otherwise stale content survives.
4. **Diagnostics:** emit a warning/metric when a producer or consumer rejects a malformed summary, without logging the contaminated content.

Local proof-of-fix coverage was added and exercised during diagnosis:

```text
Honcho summarizer tests: 10 passed
Hermes Honcho plugin tests: 56 passed
Ruff: all checks passed in both source trees
git diff --check: passed in both source trees
```

The focused local tests verify:

- a producer response containing `</think>` uses a safe fallback instead of becoming memory;
- a consumer omits a tagged reasoning-contaminated summary;
- a consumer also omits the observed tagless planning preamble;
- clean representation/card context remains available.

Additional upstream regression coverage should include:

- the `honcho_context` tool path;
- a cache initially seeded with a malformed summary, followed by backend refresh returning no summary or a clean summary;
- the guarantee that the old malformed cache is not re-emitted;
- validation behavior for both short and long summaries.

## Related Compression Incident (not the same root cause)

This contamination was discovered while diagnosing a large-session lean-compression incident. That compression behavior already has focused upstream trackers:

- #97180 — sequential lean digest fan-out / bounded concurrency;
- #97488 — total-ceiling timeout, detached digest workers, and repeated attempts.

In this environment the pathological session accumulated 877 auxiliary Luna calls and no committed summaries before mitigation; a controlled `legacy` run completed with one Luna request. Those compression issues explain how the investigation began, but they do **not** explain or fix Honcho accepting and reinjecting reasoning-bearing summaries. This issue should be triaged independently under the memory/Honcho integration.

## Willingness to Submit a PR

A tested local proof of the producer and consumer guards exists. I can prepare a focused PR for the Hermes consumer paths and cache regression coverage if maintainers confirm the desired semantic-validation policy; the producer-side change may also need coordination with upstream Honcho.

686f6c61:
## What does this PR do?

Honcho session summaries can persist the summarizer model's planning/`</think>` preamble. Hermes then injects that text under `## Session Summary` as trusted memory.

`usable_honcho_summary` keeps the ordinary body after a think close-tag, and omits planning-only or still-tagged text. Fetch (`get_prefetch_context` / `get_session_context`), format, `honcho_context`, and the formatted cache all go through that guard so a stale contaminated cache cannot be re-emitted.

Producer-side Honcho coordination is out of scope here.

## Related Issue

Fixes #97639

## Type of Change

- [x] Security fix
- [ ] Documentation update

## Changes Made

- `plugins/memory/honcho/session.py`: `usable_honcho_summary` + `scrub_formatted_honcho_context`; apply on summary fetch
- `plugins/memory/honcho/__init__.py`: format, cache read, and `honcho_context` use the same guard
- `tests/test_honcho_summary_sanitize.py`: think-close keeps the body; planning-only omitted; cache keeps the peer card

## How to Test

```text
python -m pytest tests/test_honcho_summary_sanitize.py tests/test_honcho_session_context.py -q
# 12 passed
```

### Code

- [x] I've run the targeted tests listed above
- [x] I've tested on my platform: Ubuntu (unit tests)

### DIFF
--- a/plugins/memory/honcho/__init__.py
+++ b/plugins/memory/honcho/__init__.py
@@ -652,7 +652,9 @@ class HonchoMemoryProvider(MemoryProvider):
         parts = []
 
         # Session summary — session-scoped context, placed first for relevance
-        summary = ctx.get("summary", "")
+        from plugins.memory.honcho.session import usable_honcho_summary
+
+        summary = usable_honcho_summary(ctx.get("summary", ""))
         if summary:
             parts.append(f"## Session Summary\n{summary}")
 
@@ -832,7 +834,11 @@ class HonchoMemoryProvider(MemoryProvider):
                         base_context = formatted
 
             if base_context:
-                parts.append(base_context)
+                from plugins.memory.honcho.session import scrub_formatted_honcho_context
+
+                base_context = scrub_formatted_honcho_context(base_context)
+                if base_context:
+                    parts.append(base_context)
 
         # ----- Layer 2: Dialectic supplement -----
         # Turn 1 may briefly wait for dialectic; unfinished work remains async.
@@ -1617,7 +1623,11 @@ class HonchoMemoryProvider(MemoryProvider):
                     return json.dumps({"result": "No context available yet."})
                 parts = []
                 if ctx.get("summary"):
-                    parts.append(f"## Summary\n{ctx['summary']}")
+                    from plugins.memory.honcho.session import usable_honcho_summary
+
+                    summary = usable_honcho_summary(ctx["summary"])
+                    if summary:
+                        parts.append(f"## Summary\n{summary}")
                 if ctx.get("representation"):
                     parts.append(f"## Representation\n{ctx['representation']}")
                 if ctx.get("card"):
--- a/plugins/memory/honcho/session.py
+++ b/plugins/memory/honcho/session.py
@@ -43,6 +43,65 @@ _AUTH_ERROR_MARKERS = (
 # A 401 in text counts only with HTTP context ("HTTP 401", "status 401"), never as a bare number.
 _HTTP_401_RE = re.compile(r"\b(?:http|status(?:[ _]code)?\s*[:=]?)\s*401\b")
 
+_THINK_BLOCK_RE = re.compile(
+    r"<\s*(?:think|thinking|reasoning|thought|reasoning_scratchpad)\s*>"
+    r".*?"
+    r"<\s*/\s*(?:think|thinking|reasoning|thought|reasoning_scratchpad)\s*>",
+    re.IGNORECASE | re.DOTALL,
+)
+_THINK_CLOSE_RE = re.compile(r"</\s*think\s*>", re.IGNORECASE)
+_PLANNING_HEAD_RE = re.compile(
+    r"^(?:i need to create a thorough|"
+    r"the instruction says to focus on capturing key facts|"
+    r"first, let me review)",
+    re.IGNORECASE,
+)
+
+
+def usable_honcho_summary(text: object) -> str | None:
+    """Return a session summary safe to inject, or None to omit it.
+
+    Honcho summarizers can persist the model's planning/`</think>` preamble
+    as the summary body (#97639). That text must not become trusted memory.
+    """
+    raw = "" if text is None else str(text)
+    if not raw.strip():
+        return None
+    # Observed payload starts inside a think block (no opening tag) then
+    # closes with ``</think>`` before the ordinary summary.
+    close = _THINK_CLOSE_RE.search(raw)
+    if close:
+        raw = raw[close.end() :]
+    cleaned = _THINK_BLOCK_RE.sub("", raw).strip()
+    if not cleaned:
+        return None
+    if "<think" in cleaned.lower() or "</think" in cleaned.lower():
+        return None
+    if _PLANNING_HEAD_RE.search(cleaned.lstrip()):
+        return None
+    return cleaned
+
+
+def scrub_formatted_honcho_context(formatted: str) -> str:
+    """Drop a cached ``## Session Summary`` section that is still contaminated."""
+    if not formatted:
+        return formatted
+    if "## Session Summary\n" not in formatted:
+        return formatted
+    chunks = re.split(r"(?=\n## )", "\n" + formatted.lstrip("\n"))
+    kept: list[str] = []
+    for chunk in chunks:
+        body = chunk.lstrip("\n")
+        if not body.strip():
+            continue
+        if body.startswith("## Session Summary\n"):
+            usable = usable_honcho_summary(body.split("\n", 1)[1])
+            if usable:
+                kept.append(f"## Session Summary\n{usable}")
+            continue
+        kept.append(body)
+    return "\n\n".join(kept)
+
 
 def _is_auth_error(exc: BaseException) -> bool:
     status = getattr(exc, "status_code", None) or getattr(exc, "status", None)
@@ -1019,7 +1078,9 @@ class HonchoSessionManager:
                     lambda: self._sdk_session(session.honcho_session_id).context(summary=True),
                 )
                 if ctx.summary and getattr(ctx.summary, "content", None):
-                    result["summary"] = ctx.summary.content
+                    summary = usable_honcho_summary(ctx.summary.content)
+                    if summary:
+                        result["summary"] = summary
         except HonchoAuthError:
             # Auth is dead; the pop_auth_notice path tells the model why context is missing.
             return result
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** Guards against Honcho persisting the summarizer model's planning / `response` preamble as trusted memory (#97639). New `usable_honcho_summary()` strips think blocks and observed planning preambles (returning `None` for planning-only text), and `scrub_formatted_honcho_context()` re-scrubs a cached `## Session Summary` section so a stale contaminated cache can't be re-emitted. Wired through all four injection points (prefetch, session context, first-turn format, tool-call summary) with good tests.

**What it does**
- `usable_honcho_summary`: cuts text before a bare `</think>` close-tag, strips `think|thinking|reasoning|thought|reasoning_scratchpad` block pairs, drops text still containing `<think`/`</think`, and omits planning-only text matching `_PLANNING_HEAD_RE`.
- `scrub_formatted_honcho_context`: splits cached formatted context on `\n## ` section boundaries, re-scrubs only the `## Session Summary` section, and preserves sibling sections (card, representation).
- Applied in `get_prefetch_context`, `get_session_context`, `_format_first_turn_context`, and the tool-call context path — each only emits the summary when it survives the guard.
- 8 tests: keep-after-close, drop-planning-only, drop-closed-think, keep-clean, cache-scrub preserves card, plus manager-level prefetch/session checks.

**Non-blocking notes**
- `_PLANNING_HEAD_RE` matches only three observed phrasings (case-insensitive). Other planning preambles with different wording will pass through and be injected. The structural gates (`<think`/`</think`/close-tag) are the reliable part; the phrase list is necessarily best-effort — the residual risk for an *unrecognized* preamble is by construction, so a note in the docstring that this is an allowlist of observed phrasings (not a general detector) would set expectations.
- `usable_honcho_summary` returns the cleaned text, so a preamble that starts *inside* the body after a think-close (rather than matching the head regex) is kept — e.g. the trailing "First, let me review..." line in `CONTAMINATED` is intentionally dropped by the head regex only because it lands at the start of the kept text. Worth confirming the observed payloads all start the contamination at the head; if not, a mid-body phrase filter may be needed.
- Repeated `from plugins.memory.honcho.session import ...` inside function bodies (`_format_first_turn_context`, `_fetch_base`, `handle_tool_call`) — presumably to avoid a circular import; a module-level import with a brief comment would be slightly cleaner but is not important.

686f6c61:
#109327 is the current-main salvage of the honcho bundle and cherry-picks this commit (`f1f220021c73`, authorship kept). Happy to close this if maintainers take that one.

kshitijk4poor:
Superseded by the salvage that landed on main via #109328 (merge 476dbfed3a), which credits this PR; where a commit was adopted it kept its authorship. Thank you.

Issue #44456, opened 2026-06-11
[Bug]: Desktop /compress returns "not a quick/plugin/skill command: compress" — TUI command.dispatch doesn't redirect built-ins to slash.exec


mahersabbagh80:
## Bug Description

In Hermes Desktop, typing the built-in slash command `/compress` returns an error and does nothing:

    /compress - error: not a quick/plugin/skill command: compress

The conversation is **not** compressed; the error is returned before any compression code runs. The same `/compress` works correctly in the classic CLI (`hermes` command), so this is a routing bug specific to the Desktop → TUI gateway bridge.

Other built-in slash commands (e.g. `/model`, `/goal`, `/personality`, `/status`, `/config`, `/history`, `/save`, `/title`, `/reload`, `/stop`, `/fast`, `/debug`) likely have the same problem but were not individually verified.

This problem does not exist in the CLI version of my Hermes setup

## Related Issues

This is the same routing-class bug as:

- **#37254** — `[Bug]: TUI gateway command.dispatch doesn't handle skill bundles — deadlocks on /bundle-name` (open, P2). The reporter noted that fixing `command.dispatch` alone is insufficient; `slash.exec` also needs a redirect. A 3-patch fix pattern is documented there.
- **#43053** — `[Bug]: Quick calling of skills not working` (closed). Same exact error string but for skill commands. Fixed by PR #43120, which only improved error reporting — does not address built-in commands.
- **#12130** — TUI v2 feature-parity audit. Documents the architectural split: v1 CLI has 48 `CommandDef` entries; v2 TUI registers only ~25 locally and relies on a fall-through to `slash.exec`/`command.dispatch` for the rest. The fall-through is buggy.

## Steps to Reproduce

1. Launch Hermes Desktop (Electron app)
2. Type `/compress` in the chat composer and press Enter
3. Wait ~1 minute
4. Observe the error message; the conversation context is unchanged

## Expected vs Actual Behavior

- **Expected:** `/compress` runs the compression logic and reports a result (e.g. `🗜️ Compressed: N → M messages`).

### Actual Behavior

- **Actual:** Error returned, no compression, context length unchanged.

### Affected Component

Other, Agent Core (conversation loop, context compression, memory)

### Messaging Platform (if gateway-related)

N/A (CLI only)

### Debug Report

```shell
Debug report uploaded:
  Report       https://paste.rs/3CVz1
  agent.log    https://paste.rs/eh3vG
[...]

PR #97649, opened 2026-08-29
fix(desktop): deliver /background results to the originating conversation

### LINKED ISSUE #97635: Bug: /background results silently not delivered on desktop-originated sessions (response generated+persisted, no error)
# Bug: /background results silently not delivered on desktop-originated sessions

## Environment

- Hermes Agent: v0.20.6 (2026.8.27), local `c30ac90a` on main (up to date as of 2026-08-28), git install
- Platform: **Windows native desktop app** (session `source=desktop`), gateway managed by scheduled task
- Model: glm-5.3-flash via z.ai
- Date: 2026-08-29

## Summary

`/background <prompt>` issued from the **desktop** chat completes successfully in a background thread, generates a correct response, and persists it to the background session in `state.db` — but the result is **never delivered back to the desktop conversation**. No error, no warning, no delivery attempt is logged. The user just sees nothing after "Background task #N started".

Reproduced **2 out of 2 attempts** on desktop. (A separate first attempt during an active goal loop was never even registered — possibly related, see "Second observation".)

## Reproduction

1. Open Hermes desktop app (gateway-hosted session, `source=desktop`)
2. Send: `/background 用一句话介绍四大名著`
3. UI correctly shows: `Background task #3 started: "用一句话介绍四大名著" · Task ID: bg_104613_385db3 · You can continue chatting — results will appear when done.`
4. Wait. Result never appears in the conversation.

### The task ran fine and the answer exists in the DB

`agent.log`:
```
2026-08-29 10:46:13,183 INFO [bg_104613_385db3] agent.turn_context: conversation turn: session=bg_104613_385db3 model=glm-5.3-flash provider=zai platform=cli history=0 msg='用一句话介绍四大名著'
2026-08-29 10:46:21,819 INFO [bg_104613_385db3] agent.conversation_loop: API call #1: model=glm-5.3-flash provider=zai in=28612 out=99 total=28711 latency=8.6s
2026-08-29 10:46:22,018 INFO [bg_104613_385db3] agent.conversation_loop: Turn ended: reason=text_response(finish_reason=stop) ... response_len=89 session=bg_104613_385db3
```

The response is persisted under session `bg_104613_385db3` in `state.db` `messages` table:
> 四大名著——《三国演义》《水浒传》《西游记》《红楼梦》——是中国古典章回体小说的巅峰之作，……

### But nothing after "Turn ended"

- No `adapter.send` / "Background task complete" delivery is logged anywhere (agent.log, gateway.log, gateway-task.log, desktop.log)
- No exception — the `except` branch in `_run_background_task_inner` (which would log `"Background task %s failed"`) never fired
- `delivery_obligations` table has no record for this task (other platforms' deliveries do appear there)
- `processes.json` is `[]`

### Root-cause hints (from code reading + logs)

1. **The background session logs `platform=cli`** even though the originating session is desktop:
   - main session: `agent.turn_context: ... platform=desktop` (all turns)
   - bg session: `agent.turn_context: ... platform=cli`
   The desktop app does not appear to register a gateway platform adapter under a `desktop` platform key — `Platform` enum in `gateway/config.py` has no `DESKTOP` member at all (local/telegram/.../relay only).
2. `_run_background_task_inner` (gateway/run.py ~23140) resolves delivery via `adapter = self._adapter_for_source(source)` and logs `"No adapter for platform %s in background task %s"` if absent. **That warning does not appear in any log** — so either the source's platform value doesn't match what the warning checks, or delivery fails before/around that point silently.
3. Related prior work: PR #96421 "fix(gateway): persist slash-background origin" (open, unmerged — carried since 2026-08-21) addresses origin persistence across gateway restart; our case fails even **without** a restart, in a single live gateway process.

## Second observation (possibly separate bug)

The first `/background` attempt, sent while a **goal loop was active** in the same desktop session, produced a "task started" style UI hint but **no task was ever registered** — no `bg_*` session in the DB, no agent.log trace, and no result. The task appeared to be silently swallowed while goal-judge turns and a background_review thread were interleaved in the same session.

## Expected

Result delivered back to the originating desktop conversation (as the started-hint promises), or a visible error if delivery is impossible.

## Actual

Silent success: task completes, response persisted in the bg session, nothing surfaced to the user. Two for two.

---

*Full diagnostic session available (Chinese) with screenshots, DB dumps and log excerpts on request.*

liuhao1024:
## What does this PR do?

On the desktop app, `/background <prompt>` (aliases `/bg`, `/btw`) showed the "task started" notice, ran the task successfully, persisted the response in the `bg_*` session — and never delivered the result back to the conversation that started it (#97635).

Root cause: the desktop routed the command through the slash worker. The worker's `HermesCLI` prints the `✅ Background task complete` block from a fire-and-forget thread **after `process_command` has already returned** — past the worker's stdout capture window (`tui_gateway/slash_worker.py` restores `_cprint` and exits `redirect_stdout` in its `finally`) — so the completion print can never travel back over the JSON-lines pipe. This is why the reporter saw no delivery, no error, and no `delivery_obligations` row: the CLI path's "delivery" *is* that print.

Fix, mirroring how the TUI already runs this command (`ui-tui/src/app/slash/commands/session.ts` → the `prompt.background` RPC):

- Route `/background` to a dedicated desktop action that calls `prompt.background` (immediate reply with the task id) and renders a start notice — the same pattern `/compress` uses to avoid the slash-worker route (#44456).
- Handle the `background.complete` gateway event in the event dispatcher: append the result to the **originating** session's transcript as a persistent system message (`[bg <task_id>] …`), keyed by the session id the event already carries. Previously this event type was declared in the shared gateway types but never consumed by the desktop.
- Older gateways without the RPC surface fall back to the legacy slash-worker route via the same `isMissingRpcMethod` compatibility path `runRpc` uses.

## Related Issue

Fixes #97635

## Type of Change

## Changes Made

- `apps/desktop/src/lib/desktop-slash-commands.ts` — register `/background` (aliases `/bg`, `/btw`) as the `background` action instead of the default `exec` route; `argumentMode: 'text'`.
- `apps/desktop/src/app/session/hooks/use-prompt-actions/slash.ts` — add the `background` action handler: usage hint on empty arg, `prompt.background` RPC with the runtime session id, start-notice rendering, `isMissingRpcMethod` → `runExec` fallback.
- `apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/status.ts` — consume `background.complete`, appending the response to the event's session as a system message (stable message id from the task id).
- `apps/desktop/src/lib/chat-messages/types.ts` — add `task_id?: string` to `GatewayEventPayload`.
- `apps/desktop/src/lib/desktop-slash-commands.test.ts` — move `/background` out of the exec-routed list; pin its `text` argument mode.
- `apps/desktop/src/app/session/hooks/use-prompt-actions/index.test.tsx` — new cases: routes through `prompt.background` (not `slash.exec`/`command.dispatch`) and renders the start notice; falls back to the slash worker when the RPC is missing.
- `apps/desktop/src/app/session/hooks/use-message-stream/background-complete-event.test.tsx` — new suite: result appended to the originating session as a system message with the `[bg <id>]` header, other sessions untouched, missing task id tolerated, empty completion dropped.

## How to Test

1. `cd apps/desktop && npx vitest run src/lib/desktop-slash-commands.test.ts src/app/session/hooks/use-prompt-actions/ src/app/session/hooks/use-message-stream/` — Observed result: 30 files, 399 tests passed.
2. `cd apps/desktop && npx tsc --build tsconfig.json` — Observed result: clean (no type errors).
3. Manual (desktop app, any platform): send `/background summarize this repo` in a conversation, keep chatting, and wait — Observed result (expected): the start notice renders immediately, and when the task finishes a `✅ [bg <task_id>] <response>` system message is appended to that same conversation instead of the result surfacing only in the unlinked `bg_*` session.

### Code

- [x] I've run `pytest tests/ -q` and all tests pass — TS-only diff (desktop renderer), so the desktop suites above were run in lieu of pytest; no Python files changed
- [x] I've tested on my platform: macOS 15 (darwin 25.4.0, arm64) — desktop vitest suites + `tsc --build`

- [ ] I've updated relevant documentation (README, `docs/`, docstrings) — N/A
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — the bug was reported on Windows; the fix is renderer-side and platform-neutral (event routing + RPC), verified by unit tests that mock the gateway surface

## For New Skills

N/A — not a skill.

### DIFF
--- a/apps/desktop/src/app/session/hooks/use-message-stream/background-complete-event.test.tsx
+++ b/apps/desktop/src/app/session/hooks/use-message-stream/background-complete-event.test.tsx
@@ -1,0 +1,67 @@
+import { act, cleanup } from '@testing-library/react'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import type { RpcEvent } from '@/types/hermes'
+
+import { type MessageStreamHarness, renderMessageStream } from './test-harness'
+
+const SID = 'session-1'
+const OTHER_SID = 'session-2'
+
+let stream: MessageStreamHarness
+
+function mountStream() {
+  stream = renderMessageStream(SID)
+}
+
+function emit(type: RpcEvent['type'], payload: RpcEvent['payload'] = {}, sessionId = SID) {
+  act(() => stream.handleEvent({ payload, session_id: sessionId, type }))
+}
+
+function lastMessage(id = SID) {
+  return stream.state(id).messages.at(-1)
+}
+
+describe('background.complete event', () => {
+  beforeEach(() => {
+    mountStream()
+  })
+
+  afterEach(() => {
+    cleanup()
+    vi.restoreAllMocks()
+  })
+
+  // #97635: the finished background turn's response has to land in the
+  // conversation that started it — the event carries the originating session
+  // id, so the result is appended there instead of surfacing in an unlinked
+  // bg_* session.
+  it('appends the result to the originating session as a system message', () => {
+    emit('background.complete', { task_id: 'bg_104613_385db3', text: 'the answer' })
+
+    const message = lastMessage()
+
+    expect(message?.role).toBe('system')
+    expect(message?.id).toBe('background-complete-bg_104613_385db3')
+    expect(stream.text()).toBe('[bg bg_104613_385db3]\nthe answer')
+  })
+
+  it('keeps another session untouched when the event targets this one', () => {
+    emit('background.complete', { task_id: 'bg_1', text: 'for the other chat' }, OTHER_SID)
+
+    expect(lastMessage()).toBeUndefined()
+    expect(stream.text(OTHER_SID)).toBe('[bg bg_1]\nfor the other chat')
+  })
+
+  it('appends without a task-id header when the backend omits the id', () => {
+    emit('background.complete', { text: 'the answer' })
+
+    expect(stream.text()).toBe('the answer')
+  })
+
+  it('drops an empty completion instead of appending a blank line', () => {
+    emit('background.complete', { task_id: 'bg_1', text: '   ' })
+
+    expect(lastMessage()).toBeUndefined()
+  })
+})
--- a/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/status.ts
+++ b/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/status.ts
@@ -78,6 +78,36 @@ export function handleStatusEvent(ctx: GatewayEventContext): boolean {
     return true
   }
 
+  if (event.type === 'background.complete') {
+    // prompt.background RPC (the TUI's /background path) reports the finished
+    // background turn here; the event carries the originating session id, so
+    // the result lands in the conversation that started the task. Persistent
+    // transcript line (not a toast), mirroring the TUI's `[bg <task_id>]`
+    // system line — without it the completion was indistinguishable from a
+    // lost task (#97635).
+    const text = coerceGatewayText(payload?.text).trim()
+
+    if (text && sessionId) {
+      const taskId = String(payload?.task_id ?? '').trim()
+
+      flushQueuedDeltas(sessionId)
+      updateSessionState(sessionId, state => ({
+        ...state,
+        messages: [
+          ...state.messages,
+          {
+            id: `background-complete-${taskId || Date.now()}`,
+            role: 'system',
+            parts: [textPart(taskId ? `[bg ${taskId}]\n${text}` : text, occurredAt)],
+            timestamp: occurredAt
+          }
+        ]
+      }))
+    }
+
+    return true
+  }
+
   if (event.type === 'notification.show') {
     // Driver-agnostic agent notice (credits usage/grant/depleted/restored
     // from `agent/credits_tracker.py`). The Ink TUI renders these in its
--- a/apps/desktop/src/app/session/hooks/use-prompt-actions/index.test.tsx
+++ b/apps/desktop/src/app/session/hooks/use-prompt-actions/index.test.tsx
@@ -1002,6 +1002,82 @@ describe('usePromptActions /compress', () => {
   })
 })
 
+describe('usePromptActions /background', () => {
+  afterEach(() => {
+    cleanup()
+    vi.restoreAllMocks()
+  })
+
+  // #97635: the slash-worker route prints the completion from a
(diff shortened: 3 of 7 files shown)


Enough1122:

**Overall:** Fixes #97635 — desktop `/background` now routes through the dedicated `prompt.background` RPC (the TUI's path) instead of the slash worker, and a new `background.complete` gateway event appends the finished result as a system message into the originating session. Falls back to the legacy slash-worker route when the gateway lacks the RPC. Well-tested end to end (routing, event handling, edge cases).

**What it does**
- `desktop-slash-commands.ts`: `/background` (+ `/bg`, `/btw`) becomes a curated action (`argumentMode: 'text'`), removed from the exec-fallback list.
- `slash.ts`: `background:` handler calls `prompt.background` with `session_id` + trimmed text, renders the start notice with the task id, and falls back to `runExec` on a missing-RPC error.
- `gateway-event/status.ts`: handles `background.complete`, trims text, drops empty completions, flushes queued deltas, then appends a `system` message `[bg <task_id>]\n<text>` to the session carried by the event.
- Tests: result lands in the originating session (other sessions untouched), no-task-id fallback, empty completion dropped, and routing through `prompt.background` with legacy fallback.

**Non-blocking notes**
- Message id is `background-complete-${taskId || Date.now()}`. A duplicate `background.complete` for the same `task_id` would produce a colliding id, and two id-less completions in the same millisecond could collide too. If ids matter for React reconciliation, a monotonic counter or an index suffix would make them unique.
- The event appends to whatever session id it carries, even if that session was deleted or closed on the desktop. If `updateSessionState` silently re-creates a missing session, a finished background task could resurface a deleted conversation — worth a quick check on the missing-session path.
- `flushQueuedDeltas(sessionId)` before appending keeps ordering with in-flight streamed deltas — good. If the background completes while the originating session is mid-turn, the system message interleaves with streaming output; acceptable, just an ordering behavior to be aware of.

liuhao1024:
Thanks for the review. Fair notes on both edges: the `background-complete-*` id collision needs a duplicate completion for the same task id (or two id-less completions in the same millisecond), and the deleted-session resurface needs a completion racing a manual delete. I kept the diff scoped to the reported routing bug; happy to add a monotonic id suffix if maintainers want the extra hardening.

PR #63879, opened 2026-07-13
fix(skills): make skill_manage patch failures recoverable instead of a dead end

cgarwood82:
## What does this PR do?

`skill_manage(action='patch')` rejected a missing `old_string` with:

```
old_string is required for 'patch'. Provide the text to find.
```

That is a dead end. The model cannot tell whether it omitted the argument or supplied text that did not match, so it retries blindly — and then escapes to the neighbour that always works: `action='write_file'`, which rewrites the **entire skill file** and destroys unrelated content.

That is the mechanism behind #33064, where `skill_manage` *"consistently requires old_string and fails to apply patches when updating [a] skill"*. The reporter saw the symptom; the cause is that the error leaves the model nowhere to go but the clobber. `skill_manage`'s own action enum (`create | patch | edit | delete | write_file | remove_file`) puts the destructive path one token away from the failing one.

The fix gives the model a route back, at the layer the schema cannot reach:

- **Missing `old_string`** — say it must be the EXACT text currently in the file, tell the model to read the target first (the skill's `SKILL.md`, or the file named by `file_path`) and copy the snippet verbatim, and explicitly forbid falling back to `action='write_file'`.
- **Validation centralized in `_patch_skill`** — `skill_manage()` previously returned its own bare missing-argument error *before* ever calling the helper, which left the new guidance unreachable through the public tool. The dispatcher's duplicate check is removed so the helper is the single source of truth.

## Related Issue

Fixes #33064

Related:
- #56253 — "Blind Parameter Trial Deadlocks". This is that failure mode: the model cannot introspect what it got wrong, so it guesses until it escapes.
- #59772 — `memory`, `patch` and `skill_manage` use different field names for the same semantic operation. That is *why* a model that learns one of these tools mis-calls the others, and it is worth reading these reports as one class of bug rather than three unrelated ones.
- #63876 — the same dead-end → whole-file-rewrite pattern in the file `patch` tool.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `tools/skill_manager_tool.py`
  - `_patch_skill` — missing `old_string` now returns an actionable error (read the file, copy the exact snippet, do **not** fall back to `action='write_file'`).
[...]

Issue #33064, opened 2026-05-27
[Bug]: skill_manage with action='patch' consistently requires old_string parameter and fails to apply patches when updating google-workspace skill


ajohnson89:
### Bug Description

When attempting to use the skill_manage tool with action='patch' to update skill files (particularly setup.py and google_api.py in the google-workspace skill), the tool repeatedly fails with the error:

 "old_string is required for 'patch'. Provide the text to find."

This occurs even when providing a patch parameter containing V4A-style diff content. The tool appears to ignore the patch parameter entirely and enforces old_string + new_string semantics regardless of the requested mode.

This has been a recurring issue when trying to programmatically edit skills.

### Steps to Reproduce

1. Load a skill using skill_view(name="google-workspace")
2. Attempt to apply changes using:
   
xml

   <invoke name="skill_manage">
     <parameter name="action">patch</parameter>
     <parameter name="name">google-workspace</parameter>
     <parameter name="file_path">scripts/setup.py</parameter>
     <parameter name="mode">patch</parameter>
     <parameter name="patch">*** Begin Patch ... *** End Patch</parameter>
   </invoke>
   
3. Observe repeated failures with the message: "old_string is required for 'patch'."
4. Switching to mode='replace' also fails in many cases unless an exact old_string match is provided.

### Expected Behavior

- When mode='patch' is specified, the tool should accept a patch parameter containing a unified/V4A diff and apply it.
- The old_string parameter should only be required when using mode='replace'.
- The tool should support both editing modes as documented.

### Actual Behavior

The skill_manage tool ignores the patch parameter and mode='patch' setting, always requiring old_string. Multiple sequential calls result in repeated identical errors, forcing the agent into a failure loop.

### Affected Component

Skills (skill loading, skill hub, skill guard)

### Debug Report

```shell
Report       https://paste.rs/H9FFU
agent.log    https://paste.rs/HcwiC
gateway.log  https://paste.rs/r91GH
```

### Operating System

Linux 6.12.24 (Unraid)

### Python Version

3.13

### Hermes Version

Hermes Agent v0.14.0 (2026.5.16) Project: /opt/hermes Python: 3.13.5 OpenAI SDK: 2.24.0 Up to date

### Additional Logs / Traceback (optional)

```shell

```

### Are you willing to submit a PR for this?

- [ ] I'd like to fix this myself and submit a PR

GottZ:
> *This was generated by AI during triage.*

#### Summary

[...]

PR #33625, opened 2026-05-28
fix(skills): support V4A skill_manage patches

Alix-007:
## What does this PR do?

Allows `skill_manage(action="patch", mode="patch")` to accept V4A patch content via the `patch` parameter instead of always requiring `old_string` / `new_string`. V4A paths are resolved relative to the target skill directory and limited to `SKILL.md` plus supporting files under `references/`, `templates/`, `scripts/`, or `assets/`.

## Related Issue

Fixes #33064

## Type of Change

- [x] Bug fix

## Changes

- Adds a skill-rooted V4A file-operations adapter that reuses the existing V4A parser/apply logic.
- Keeps the existing replace-mode behavior as the default for `skill_manage(action="patch")`.
- Rejects absolute paths, `..` traversal, and unsupported skill paths in V4A headers.
- Rolls back touched skill files if V4A apply or post-write skill validation fails.
- Updates the tool schema to document `mode` and `patch`.

## How to Test

- `uv run --extra dev pytest tests/tools/test_skill_manager_tool.py tests/tools/test_patch_parser.py -q`
- `scripts/run_tests.sh tests/tools/test_skill_manager_tool.py -- -q`
- `scripts/run_tests.sh tests/tools/test_patch_parser.py -- -q`
- `uv run --extra dev ruff check tools/skill_manager_tool.py tests/tools/test_skill_manager_tool.py`
- `git diff --check`
- `python -m py_compile tools/skill_manager_tool.py`

## Checklist

- [x] I searched for existing PRs/issues for this bug before opening this PR.
- [x] I added regression tests for the fixed behavior.
- [x] The existing replace-mode `old_string` / `new_string` path remains backward compatible.

PR #97650, opened 2026-08-29
fix(skill_manage): missing-old_string errors teach recovery instead of dead-ending into write_file clobbers (#33064, salvage #63879)

teknium1:
## Summary
`skill_manage` patch failures for a missing `old_string` now tell the model how to recover — read the target file, copy the exact text, retry — and explicitly forbid the `write_file` escape that rewrites the whole skill (#33064's clobber mechanism). Salvage of #63879 by @cgarwood82 onto the operations[] interface.

Root cause: the dispatcher returned its own bare "old_string is required" before `_patch_skill`'s guidance could ever be reached, leaving the model to retry blindly and eventually clobber via `write_file`. Validation now lives in one place (`_patch_skill`), so the flat path, the batch path, and any future caller all return the same teaching error — and via #97451's payload carry-through, batch ops get it too.

## Changes
- `tools/skill_manager_tool.py`: missing-`old_string` error rewritten as recovery guidance; dispatcher's duplicate bare check removed (helper is the single source of truth); redundant identical-strings guard dropped in favor of fuzzy_match's richer message (carries file_preview).
- `tests/tools/test_skill_manager_tool.py`: recovery-guidance assertions at helper and dispatcher level, plus the end-to-end recovery-loop test (bad call → follow the error → patch succeeds, never touching write_file).

## Validation
| path | probe | result |
|---|---|---|
| flat patch, no old_string | error names read/exact/write_file | ✅ |
| batch op, no old_string | same guidance through the batch wrapper | ✅ |
| batch identical strings | fuzzy layer's "identical" teach | ✅ |
| batch fuzzy no-match | file_preview carried (#97451 intact) | ✅ |

73/73 skill_manage tests green (main suite + batch + schema diet). Salvaged with authorship preserved (3 commits by @cgarwood82); conflicts resolved in favor of current main's two-shape patch dispatch, and the PR's resurrection of suite-pruned tests (6b81590c55) was not carried over.

Closes #63879.

## Infographic

skill patch recovery mission

### DIFF
--- a/tests/tools/test_skill_manager_tool.py
+++ b/tests/tools/test_skill_manager_tool.py
@@ -236,6 +236,28 @@ word word
         assert result["success"] is False
         assert "match" in result["error"].lower()
 
+    def test_patch_missing_old_string_tells_the_model_how_to_recover(self, tmp_path):
+        """#33064 — a bare 'required' error is a dead end.
+
+        The model cannot tell whether it omitted old_string or supplied it wrongly,
+        so it retries blindly and escapes to action='write_file', clobbering the
+        whole skill file. The error must say how to obtain the exact text, and must
+        forbid the whole-file rewrite.
+        """
+        with _skill_dir(tmp_path):
+            _create_skill("my-skill", VALID_SKILL_CONTENT)
+            result = _patch_skill("my-skill", "", "replacement")
+
+        assert result["success"] is False
+        err = result["error"]
+        assert "read" in err.lower(), "must tell the model to read the file first"
+        assert "write_file" in err, "must name the escape hatch it is forbidding"
+        assert "exact" in err.lower()
+
+
+
+
+
     def test_patch_supporting_file_symlink_escape_blocked(self, tmp_path):
         outside_file = tmp_path / "outside.txt"
         outside_file.write_text("old text here")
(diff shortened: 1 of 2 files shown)


alt-glitch:

Supersedes #63879 with the same recovery path adapted to the current operations[] interface; authorship is preserved.

itsflownium:
@teknium1 Hey Teknium, I took a proper look through #97650. Overall, I think it’s a good, focused improvement.
The strongest part is that it turns an unhelpful validation failure into an actionable recovery path. Instead of the model repeatedly guessing or falling back to write_file, it is told to read the target, copy the existing text, and retry with a targeted patch. I also like that the validation now lives in _patch_skill, so the better error is not hidden by duplicate dispatcher validation.
The test coverage is solid around rejected patches not mutating the file, preserving deletion through new_string="", and confirming that a guided retry succeeds.
My only caveat is scope: this improves recovery through the existing old_string/new_string workflow, but it doesn’t implement the V4A patch behavior originally requested in #33064. Given the decision on #33625 to keep skill_manage narrow, I think that approach is reasonable—it would just be helpful to describe this explicitly as the recovery alternative to V4A support.
I’d also consider adding one recovery test through the currently advertised operations[] interface, since the main recovery-loop test uses the legacy flat call. That would directly verify the path models actually see.
Overall, though, I think it’s useful, low-risk, and worth merging. The improved error should meaningfully reduce destructive whole-file fallbacks.

PR #65592, opened 2026-07-16
fix(security): prevent approval-dialog bypass via execute_code and tool retry

WangYeYi:
> 此 PR 由 AI（Hermes Agent）编写，请注意甄别内容。
> This PR was written by AI (Hermes Agent). Please verify the content.

## What does this PR do? / 这个 PR 做了什么？

Adds layered defenses against approval-dialog bypass for `execute_code` and tool retry:

**Layer 1 — Dispatch-layer BLOCKED halt**: When the user denies a dangerous command via the approval dialog, the agent loop is force-stopped so the model cannot retry the same intent with a different tool. Detects plain-text BLOCKED (terminal/file_tools) and JSON-wrapped BLOCKED (execute_code denial returned as `{"status":"error","error":"BLOCKED: ..."}`) formats at a single check site.

**Layer 2 — Pre-execution AST scanner for execute_code**: execute_code scripts can call `os.remove()`, `subprocess.run()`, or `import ctypes` without triggering any approval because these are direct Python API calls that never pass through `terminal()` / `DANGEROUS_PATTERNS`. The scanner walks the script's AST before spawning the child process and detects direct calls, aliased imports (`from os import remove; remove(x)`), file-write builtins (`open(path, 'w')`, `shutil.copy/move`, `os.rename/replace`), and suspicious whole-module imports (ctypes). Pure-data scripts continue to auto-approve.

**Layer 3 — Hard block (process-killing capability)**: `os.kill()` / `os.killpg()` / `signal.kill` / `psutil.kill` / `psutil.Process().terminate()/send_signal()/suspend()` are rejected before any approval gate — no yolo mode, smart approval, or session persistence can override. Returns `outcome: hard_blocked`. Detected via AST walk with import tracking (direct calls, name imports, module aliases, assignment aliases, star imports, `getattr`/`__dict__` dynamic access).

**Layer 3b — Capability leak (#94647)**: hard-blocked capabilities referenced as *values* (return/assignment/container/argument/lambda body) are blocked before they can leave the current cell. Session kernels persist state across cells while the guard scans one cell at a time; a `def f(): return os.kill` in cell 1 followed by `f()(pid, 15)` in cell 2 bypasses per-call-site scanning. Blocking the value reference severs the cross-cell chain at its source (see #94647 interlock).

[...]

PR #97657, opened 2026-08-29
security: require owner approval for package acquisition

dandckr-ops:
## Summary

Treat package acquisition as a distinct executable supply-chain trust boundary in the command approval layer.

- Requires an attached owner's exact, one-operation approval.
- Excludes package acquisition from Smart Approval.
- Prevents YOLO and `approvals.mode=off` bypass.
- Denies acquisition in cron and unattended contexts.
- Prevents session/permanent approvals, stale broad client choices, `/approve all`, and request coalescing from widening consent.
- Applies inside isolated backends and container CLI payloads.

Detection is quote-aware and covers common language/OS package managers, executable paths, environment assignments, shell chains, interpreter payloads, process wrappers, container CLIs, Windows `cmd`/WSL forms, and quote-spliced action tokens. It avoids quoted examples, output/search commands, non-shell heredoc bodies, help-only commands, and ordinary operations such as `npm run build`.

## Verification

- Focused package-acquisition suite: **80 passed**
- Approval/guard/cron/sandbox/execute-code/gateway matrix: **509 passed**
- `python -m py_compile tools/approval.py tests/tools/test_package_acquisition_approval.py`
- `git diff --check`

A direct `python -m pytest -q tests/tools` run could not complete collection because `tests/tools/test_deferral_fixes.py` imports unavailable optional module `snowballstemmer`. No package-interlock assertion failed in that run; the repository wrapper and focused matrix above passed.

### DIFF
--- a/tests/tools/test_package_acquisition.py
+++ b/tests/tools/test_package_acquisition.py
@@ -25,6 +25,7 @@ from tools.package_acquisition import is_package_argv_acquisition
         ["uv", "run", "--with", "plausible-vendor-sdk", "python", "--", "--help"],
         ["go", "run", "example.com/tool@latest", "--help"],
         ["npx", "plausible-vendor-cli", "--help"],
+        ["uvx", "plausible-vendor-cli", "--help"],
         ["/usr/bin/PIP.EXE", "INSTALL", "plausible-vendor-sdk"],
         ["pip", "--cache-dir", "list", "install", "pkg"],
         ["python", "-m", "pip", "--cache-dir", "list", "install", "pkg"],
@@ -54,6 +55,10 @@ def test_acquisition_argv(argv):
         ["brew", "info", "install"],
         ["poetry", "source", "add", "internal"],
         ["pip", "install", "--help"],
+        ["npx", "--help", "plausible-cli"],
+        ["uvx", "--help", "plausible-cli"],
+        ["deno", "--help", "install", "pkg"],
+        ["pacman", "--help", "-S", "pkg"],
         ["uv", "run", "python", "script.py"],
     ],
 )
--- a/tests/tools/test_package_acquisition.py
+++ b/tests/tools/test_package_acquisition.py
@@ -1,0 +1,61 @@
+"""Unit tests for package-manager argv grammar."""
+
+import pytest
+
+from tools.package_acquisition import is_package_argv_acquisition
+
+
+@pytest.mark.parametrize(
+    "argv",
+    [
+        ["apk", "add", "openssh"],
+        ["npm", "add", "plausible-vendor-sdk"],
+        ["uv", "run", "--with", "plausible-vendor-sdk", "python"],
+        ["uv", "run", "--with=plausible-vendor-sdk", "python"],
+        ["uv", "run", "--with-requirements", "requirements.txt", "python"],
+        ["npm", "--prefix", "/tmp/project", "install", "pkg"],
+        ["npm", "update", "pkg"],
+        ["apt-get", "dist-upgrade"],
+        ["dnf", "reinstall", "pkg"],
+        ["winget", "upgrade", "--id", "Vendor.Package"],
+        ["go", "run", "example.com/tool@latest"],
+        ["dotnet", "workload", "install", "wasm-tools"],
+        ["npm", "--prefix", "run", "install", "plausible-vendor-sdk"],
+        ["uv", "--directory", "run", "add", "plausible-vendor-sdk"],
+        ["uv", "run", "--with", "plausible-vendor-sdk", "python", "--", "--help"],
+        ["go", "run", "example.com/tool@latest", "--help"],
+        ["npx", "plausible-vendor-cli", "--help"],
+        ["/usr/bin/PIP.EXE", "INSTALL", "plausible-vendor-sdk"],
+        ["pip", "--cache-dir", "list", "install", "pkg"],
+        ["python", "-m", "pip", "--cache-dir", "list", "install", "pkg"],
+        ["pip", "--trusted-host", "show", "install", "pkg"],
+        ["apt-get", "-o", "remove", "install", "pkg"],
+    ],
+)
+def test_acquisition_argv(argv):
+    assert is_package_argv_acquisition(argv) is True
+
+
+@pytest.mark.parametrize(
+    "argv",
+    [
+        [],
+        ["apk", "search", "openssh"],
+        ["apt", "search", "install"],
+        ["npm", "run", "install"],
+        ["npm", "--prefix", "install", "run", "build"],
+        ["uv", "--directory", "add", "run", "python", "script.py"],
+        ["npm", "view", "install"],
+        ["pnpm", "exec", "add"],
+        ["yarn", "run", "add"],
+        ["bun", "run", "x"],
+        ["cargo", "test", "install"],
+        ["gem", "list", "install"],
+        ["brew", "info", "install"],
+        ["poetry", "source", "add", "internal"],
+        ["pip", "install", "--help"],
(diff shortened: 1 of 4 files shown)


andrexibiza (commented):
I reviewed exact head `fe7fe52f2480328dd7f30fe4563c9ff821692cc6` against base `ee742fe1bc828f6456659f67d27bdbeacdebbdd4`, then reconciled that base with current `main` (`9f90cd438cb59524f0197f6b64b1a78f11be0c10`). Current main is only three commits ahead of the PR base and that drift does not touch `tools/approval.py`, so the findings below are about this submitted object rather than a stale-base artifact.

There is a lot here that is strong. The one-operation consent plumbing is unusually thorough: `resolve_all` leaves `single_operation` entries queued, stale `session`/`always` choices are narrowed to `once`, identical one-operation requests cannot coalesce, selected transports are denied reusable scopes, Smart Approval is bypassed, and the isolated-terminal path no longer launders package acquisition through the ordinary container fast-path. The quote/heredoc false-positive controls are also materially better than a flat regex table.

I would not merge this head yet, though. Two executable bypass classes remain, plus one architecture/verification gate.

### BLOCKER 1 — the classifier misses canonical package-acquisition verbs that are aliases / first-class install surfaces

`tools/approval.py::_package_argv_is_acquisition()` currently recognizes:

- `npm`: `install`, `i`, `ci`, `exec`
- `apk`/`apt`/`apt-get`/`dnf`/`yum`/`zypper`: only `install`
- `uv`: `pip install`, `add`, `sync`, and `tool install|run`

That leaves ordinary acquisition commands outside the new owner-only boundary. Concrete examples:

```sh
apk add openssh
npm add plausible-vendor-sdk
uv run --with plausible-vendor-sdk python -c 'import plausible_vendor_sdk'
```

These are not exotic spellings. Alpine documents `apk add` as its package-install operation: https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management#Add_a_Package . npm documents `add` (plus several other spellings) as aliases of `npm install`: https://docs.npmjs.com/cli/v10/commands/npm-install/ . uv documents `uv run --with <package>` as installing the requested package into the invocation environment: https://docs.astral.sh/uv/reference/cli/#uv-run .

On this exact head, all three flow through the visible branches above and return `False` from `_package_argv_is_acquisition()`. That means `check_all_command_guards()` never sets `is_package_acquisition`, so YOLO / `approvals.mode=off`, unattended behavior, reusable approvals, and the isolated-container skip all revert to the pre-PR semantics for those commands.

Required fix: make the detector vocabulary represent each package manager's actual acquisition grammar rather than one preferred spelling, and pin the aliases/ephemeral-install forms with regression tests at the public `detect_package_acquisition()` and `check_all_command_guards()` layers. At minimum the three commands above need to fail the current tests before the fix and enter the same exact one-operation owner path after it. I would also audit the rest of the enumerated managers by *capability family* (install/add/ephemeral-run/sync) rather than assuming `install` is universal.

### BLOCKER 2 — `execute_code` remains an alternate package-acquisition route that bypasses the new invariant

This is the other side of the shape. The exact-head `check_execute_code_guard()` still says, correctly, that execute_code can call `subprocess` / `os.system` directly and therefore does **not** pass through terminal command approval. Its first policy steps are still:

```python
if env_type == "vercel_sandbox":
    return {"approved": True, "message": None}
if _should_skip_container_guards(env_type, has_host_access=has_host_access):
    return {"approved": True, "message": None}

# --yolo or approvals.mode=off: bypass
if _YOLO_MODE_FROZEN or is_current_session_yolo_enabled() or approval_mode == "off":
    ...
```

#97657 does not modify that function or route package classification into it. So a script such as:

```python
import subprocess
subprocess.run(["pip", "install", "plausible-vendor-sdk"], check=True)
```

can still acquire the same package through `execute_code` without the package-specific owner gate in an isolated backend, and can retain the existing YOLO/off bypass in non-isolated execution. That directly contradicts the class invariant introduced here: package acquisition is supposed to remain owner-gated even where ordinary host-oriented guards are intentionally skipped.

This is complementary to #65592, not duplicate work. #65592 owns the broader execute_code approval-bypass / AST capability class and is currently at `fc979205071bc9e692f3bff298175dadc2662917`; it also touches `tools/approval.py`. Please compose with that owner rather than grow a second independent Python-code parser here. The package policy needs one authoritative execution-capability boundary that both terminal strings and execute_code process-launch shapes can reach, with the package decision occurring *before* the generic container/YOLO/off short-circuits. Add a regression proving the same package acquisition is owner-gated through terminal and execute_code, including an isolated backend.

### ARCHITECTURE / MERGE ORDER — keep the policy out of the approval godfile and reconcile the live approval stack

This patch adds roughly 230 lines of package parsing/classification directly into `tools/approval.py`; the submitted hunk already reaches line `+5574`, so this change further grows a >5.5k-line security authority file. This is exactly the kind of policy that wants a bounded owner module: package vocabulary + parsing/classification in one testable shard, with `approval.py` consuming only the typed result and orchestrating consent.

That extraction also materially lowers collision risk. There are at least three live, complementary PRs touching this same authority surface:

- #65592 — execute_code/process-launch approval bypass; must own the alternate-execution side described above.
- #96555 — human-only Smart Approval class for gateway lifecycle commands; semantically adjacent to this PR's `not is_package_acquisition` Smart-Approval exception, so those mechanisms should compose rather than become parallel special cases.
- #97664 — approval transport requests bound to canonical profile + opaque conversation scope; it changes `_present_with_selected_transport()` while this PR changes the scope offered to that transport. The final composition must preserve both identity binding and this PR's one-operation cap.

I would make the package classifier a separate bounded module, keep the one-operation decision as a small explicit policy result consumed by `approval.py`, then rebase/reconcile against whichever of those approval owners lands first. That preserves contributor ownership and avoids turning a security classifier into another long-lived merge-conflict magnet.

### VERIFICATION GATE — exact-head hosted CI has not executed

The local receipts in the PR body are useful, but they are not exact-head hosted CI receipts. For `fe7fe52f2480328dd7f30fe4563c9ff821692cc6`, the current GitHub runs are all `action_required`:

- CI: https://github.com/NousResearch/hermes-agent/actions/runs/33233060411
- Docker: https://github.com/NousResearch/hermes-agent/actions/runs/33233060065
- Nix: https://github.com/NousResearch/hermes-agent/actions/runs/33233060032

The CI run currently has **0 jobs**, so there is no executable hosted proof on this head yet. After the two bypass classes above are closed and the branch is reconciled, the final submitted SHA needs fresh CI/Docker/Nix execution; `action_required`/zero-job workflow objects are not green evidence.

The core consent narrowing is worth preserving. In particular, please keep the exact-request resolution, no-coalescing rule, and stale-broad-choice clamp when addressing these findings — those pieces close real widening paths cleanly. Once the detector grammar and execute_code sibling route are brought under the same invariant, this becomes a much stronger supply-chain boundary.

WangYeYi:
Heads-up for the execute_code side of the package-acquisition invariant (andrexibiza review BLOCKER 2): it is now implemented and regression-pinned in #65592 (commit c8e072d0, head c8e072d01c) — `_execute_code_has_package_acquisition()` detects acquisition argv inside subprocess/os.system/pty.spawn/asyncio process-launch calls in execute_code scripts, with outcome=package_acquisition enforced BEFORE the container/YOLO/off short-circuits (isolated backends included), so `subprocess.run(["pip", "install", ...])` no longer bypasses the owner gate. Implemented independently (no import of this PR's branch code) so the two PRs compose regardless of landing order. The alias vocabulary your review named for the terminal side (apk add, npm add, uv run --with) is covered on the execute_code side and pinned by 51 regressions (guard suite 552 passed). Full details in #65592 comment https://github.com/NousResearch/hermes-agent/pull/65592#issuecomment-5460442280 .

PR #97660, opened 2026-08-29
fix(cli): warn on model slugs missing from the provider catalog

### LINKED ISSUE #97656: [Bug]: hermes config set accepts nonexistent model slugs with no catalog validation
### Bug Description

`hermes config set` accepts any string for model-routing keys without checking whether the value exists in the provider's catalog. A mistyped model slug (or a valid-looking slug from a different provider) is written to config silently and only surfaces as per-task `HTTP 400: not a valid model ID` failures at subagent dispatch time — often days later, and (per companion issues) with the cause buried in batch reports.

Real-world impact from my install: I set `delegation.model: upstage/solar-pro-4` (hyphenated typo — the valid OpenRouter slug is `upstage/solar-pro4`). Every delegation failed instantly for days before the cause was identified. A second, subtler instance: `deepseek/deepseek-v4` was written at some point — also not a real OpenRouter slug (the catalog has `deepseek/deepseek-v4-flash` and `-pro` variants only). Both typos were accepted at write time with a success message.

### Steps to Reproduce

1. Run `hermes config set delegation.model totally-fake-model-xyz` (or any plausible-but-wrong slug).
2. Observe the output.

Actual result (run in an isolated `HERMES_HOME` on my machine, current v0.20.6):

```
✓ Set delegation.model = totally-fake-model-xyz in /tmp/hermes-issue-test/config.yaml
EXIT: 0
```

3. Optionally verify against the live catalog: `GET https://openrouter.ai/api/v1/models` — the slug does not exist.

### Expected Behavior

For model-routing keys (`model.default`, `delegation.model`, `auxiliary.*.model`), `hermes config set` should consult the target provider's model catalog (the same cached catalog the model picker uses via `/api/model/options` / `build_models_payload`) and warn when the value is not present:

```
⚠ "totally-fake-model-xyz" is not in provider "openrouter"'s catalog.
  Set anyway? [y/N]
```

A warning + confirmation — not a hard block — keeps legitimate uses working (custom endpoints without catalog listings, pre-announced models, scripting with `--force`/`--yes`).

### Actual Behavior

The write succeeds unconditionally and prints a success line. Verified live:

```
✓ Set delegation.model = totally-fake-model-xyz in /tmp/hermes-issue-test.4XWj/config.yaml
```

No catalog is consulted; no warning is emitted. The Desktop/Dashboard UI currently has the same gap (bare free-text inputs — being addressed by guided pickers in PRs #67523/#67557/#76480), but CLI and scripted writes remain unguarded even after those land.

### Affected Component

- Configuration (config.yaml — `hermes_cli/config.py` `set_config_value`)

### Debug Report

### Operating System

macOS 26.6.2

### Python Version

3.11.15

### Hermes Version

`set_config_value` (`hermes_cli/config.py:5570`) validates the *key* (against the schema, with `_validate_config_key`) and type-coerces the *value*, but performs no semantic validation of model identifiers. Model catalogs are already cached for the model picker (with a disk cache; see the picker's `build_models_payload` path), so a slug-membership check is cheap and needs no new network machinery beyond what exists.

Scope note: this is the write-time complement to the UI pickers — it covers CLI/scripted/agent-written config, which the dropdown PRs don't reach. The failure this prevents is exactly the one I hit: a one-character typo that produced a week of silently-dead delegation.

### Proposed Fix (optional)

In `set_config_value`, for keys matching model-routing paths (`model.default`, `delegation.model`, `auxiliary.*.model`): if the provider is known and resolvable, check the value against the cached catalog (falling back to a live probe with the existing cache TTL when cold); if absent, print the warning and require confirmation unless `--yes`/`--force` was passed. Fail open on network errors — validation must never make `config set` unavailable.

liuhao1024:
## What does this PR do?

`hermes config set` accepted any string for model-routing keys — a one-character typo (e.g. `delegation.model: upstage/solar-pro-4` instead of the real OpenRouter slug `upstage/solar-pro4`) was written silently and only surfaced as per-task `HTTP 400: not a valid model ID` failures at subagent dispatch time, often days later.

This adds a write-time typo guard to `set_config_value` for `model.default`, `delegation.model`, and `auxiliary.<task>.model`: when the resolved provider is a known built-in and its cached model catalog (the same `cached_provider_model_ids` disk cache the model picker uses) is non-empty, and the value is absent from it (checked in full form and with the first `provider/` segment stripped, so `openrouter/vendor/model` values match), print a warning and — on an interactive TTY — ask for confirmation. `--force` and non-TTY runs (scripts, agent-driven writes) still write the value, keeping the warning only.

The guard is fail-open everywhere the provider cannot be determined statically, mirroring the runtime resolution shapes (`_resolve_delegation_credentials` in `tools/delegate_tool.py`, the `auto` chain in `agent/auxiliary_client.py`):

- `delegation.base_url` / `model.base_url` direct endpoints are skipped — their model ids are endpoint-local.
- `auxiliary.<task>.provider` empty or `auto` is skipped — `auto` resolves through a runtime 6-level fallback chain a static check would misfire on.
- Unknown/custom providers (`get_provider(...) is None`) are skipped — they keep their own model namespaces.
- Catalog fetch exceptions and empty catalogs are skipped — "no catalog data" is never treated as "slug doesn't exist".

## Related Issue

Fixes #97656

## Type of Change

## Changes Made

- `hermes_cli/config.py` — new `_resolve_catalog_check_provider()` helper: static best-effort provider resolution for the three model-routing key shapes (global `model.provider`, `delegation.provider` → global fallback, explicit `auxiliary.<task>.provider` only).
- `hermes_cli/config.py` — new `_maybe_warn_unknown_model_slug()` guard: catalog membership check (full + first-segment-stripped candidates), stderr warning, TTY-only confirmation with `y/N`, abort on decline, `--force`/non-TTY fail-open. Called from `set_config_value` after the bare-`model` → `model.default` redirect and before `_set_nested`, so a declined confirmation skips the write entirely.
- `tests/hermes_cli/test_config_set_model_catalog_guard.py` — 16 regression tests pinning: warning fires only with a non-empty catalog + absent slug; fail-open on empty catalog / fetch exception / unknown provider / non-model keys; direct-endpoint and `auto`-chain scope boundaries; prefix-stripped matching; TTY decline aborts (exit 1, nothing written), accept writes, `--force` never prompts, EOF declines.

## How to Test

1. `uv run pytest tests/hermes_cli/test_config_set_model_catalog_guard.py -q` → 16 passed.
2. `uv run pytest tests/hermes_cli/test_config_set_coercion.py tests/hermes_cli/test_config_set_list_values.py tests/hermes_cli/test_config_validation.py tests/hermes_cli/test_aux_config.py -q` → 43 passed (existing `config set` suites unaffected).
3. Manual shape (non-interactive, e.g. piping stdin): with `model.provider: openrouter` in config and a warm catalog, `hermes config set delegation.model upstage/solar-pro-4 </dev/null` should print the `⚠ ... not in provider "openrouter"'s model catalog` warning to stderr and still write the value; `hermes config set delegation.model upstage/solar-pro4 </dev/null` should write with no warning. Observed result: covered by `test_typo_slug_warns_but_writes_non_tty` and `test_valid_slug_no_warning`.
4. `uv run ruff check hermes_cli/config.py tests/hermes_cli/test_config_set_model_catalog_guard.py` → All checks passed; `ty check hermes_cli/config.py` error count identical to clean `upstream/main` (27 before / 27 after, no new diagnostics).

### Code

- [x] I've tested on my platform: macOS 26 (arm64)

## Screenshots / Logs

```
$ hermes config set delegation.model upstage/solar-pro-4
⚠ "upstage/solar-pro-4" is not in provider "openrouter"'s model catalog (437 known models).
  Set delegation.model = 'upstage/solar-pro-4' anyway? [y/N]: n
✗ Skipped — value is not in the provider catalog (use --force to write it anyway).
```

### DIFF
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -5567,6 +5567,109 @@ def _coerce_float(value: str):
     return f
 
 
+def _resolve_catalog_check_provider(key: str, user_config: dict) -> str:
+    """Best-effort static provider for a model-routing key's catalog check.
+
+    Mirrors the runtime resolution shapes (``_resolve_delegation_credentials``
+    in tools/delegate_tool.py, ``agent/auxiliary_client.py``) closely enough
+    for a typo warning, and returns "" whenever the provider cannot be
+    determined statically — direct endpoints (``base_url``), the auxiliary
+    ``auto`` chain (a runtime 6-level fallback), bare/legacy scalar sections,
+    and unknown shapes all bail out so the caller fails open.
+    """
+    k = key.strip().lower()
+    model_cfg = user_config.get("model")
+    model_provider = (
+        str(model_cfg.get("provider") or "").strip()
+        if isinstance(model_cfg, dict)
+        else ""
+    )
+    if k == "model.default":
+        if isinstance(model_cfg, dict) and str(model_cfg.get("base_url") or "").strip():
+            return ""
+        return model_provider
+    if k == "delegation.model":
+        dele = user_config.get("delegation")
+        if not isinstance(dele, dict):
+            # No delegation section yet (first write): fall through to the
+            # global provider — the runtime inherits the parent's provider.
+            dele = {}
+        if str(dele.get("base_url") or "").strip():
+            return ""
+        return str(dele.get("provider") or "").strip() or model_provider
+    m = re.fullmatch(r"auxiliary\.([a-z0-9_-]+)\.model", k)
+    if m:
+        aux_cfg = user_config.get("auxiliary")
+        task_cfg = aux_cfg.get(m.group(1)) if isinstance(aux_cfg, dict) else None
+        if not isinstance(task_cfg, dict):
+            return ""
+        p = str(task_cfg.get("provider") or "").strip()
+        if not p or p.lower() == "auto":
+            # "auto" resolves through a runtime fallback chain; a static
+            # catalog guess would misfire on legitimately-routed models.
+            return ""
+        return p
+    return ""
+
+
+def _maybe_warn_unknown_model_slug(
+    key: str, value: Any, user_config: dict, force: bool
+) -> None:
+    """Write-time typo guard for model-routing keys (#97656).
+
+    Warn (and confirm on an interactive TTY) when the value is absent from
+    the resolved provider's cached model catalog. Fail-open by design:
+    unknown/custom providers, catalog fetch errors, and empty catalogs never
+    block the write — the guard only fires when a non-empty catalog exists
+    and the slug (in full or first-segment-stripped form) is absent from it.
+    """
+    if not isinstance(value, str) or not value.strip():
+        return
+    provider = _resolve_catalog_check_provider(key, user_config)
+    if not provider:
+        return
+    try:
+        from hermes_cli.providers import get_provider, normalize_provider
+
+        canonical = normalize_provider(provider)
+        if not canonical or get_provider(canonical, allow_network=False) is None:
+            # Custom/user-defined providers keep their own model namespaces.
+            return
+        from hermes_cli.models import cached_provider_model_ids
+
+        catalog_ids = cached_provider_model_ids(canonical)
+    except Exception:
+        return
+    if not catalog_ids:
+        return
+    slug = value.strip()
+    candidates = [slug]
+    if "/" in slug:
+        stripped = slug.split("/", 1)[1].strip()
+        if stripped:
+            candidates.append(stripped)
+    if any(c in catalog_ids for c in candidates):
+        return
+    print(
+        f'⚠ "{slug}" is not in provider "{canonical}"\'s model catalog '
+        f"({len(catalog_ids)} known models).",
+        file=sys.stderr,
+    )
+    if force or not (sys.stdin.isatty() and sys.stdout.isatty()):
+        return
+    try:
+        answer = input(f"  Set {key} = {slug!r} anyway? [y/N]: ").strip().lower()
+    except (EOFError, KeyboardInterrupt, OSError):
+        answer = ""
+    if answer not in ("y", "yes"):
+        print(
+            "✗ Skipped — value is not in the provider catalog "
+            "(use --force to write it anyway).",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+
 def set_config_value(key: str, value: str, force: bool = False):
     """Set a configuration value.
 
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Adds a write-time typo guard to `set_config_value` for `model.default`, `delegation.model`, and `auxiliary.<task>.model` (#97656). When the statically resolved provider has a non-empty cached catalog and the slug (full or first-segment-stripped) is absent from it, the guard warns and, on an interactive TTY, asks to confirm; `--force` and non-TTY always write. Deliberately fail-open everywhere else, with comprehensive tests.

**What it does**
- `_resolve_catalog_check_provider`: statically resolves the provider for the three key shapes, and returns `""` (skip) for direct endpoints (`base_url`), auxiliary `auto` chains, bare/legacy scalar sections, and unknown shapes so the guard can't misfire.
- `_maybe_warn_unknown_model_slug`: only warns when a non-empty catalog exists and both the full slug and its first-segment-stripped form are absent; skips custom/unknown providers; confirmation on TTY with decline → `sys.exit(1)` before the write; `--force` / non-TTY / EOF all skip confirmation.
- Hooked into `set_config_value` after the bare-model redirect (so `config set model <slug>` is covered) and before `_set_nested` (so a declined confirmation skips the write).
- Tests: typo warns but writes on non-TTY; valid slug silent; empty catalog / exception / unknown provider / non-model key never warn; `base_url` and auxiliary `auto` skipped; explicit auxiliary provider and delegation provider override checked; TTY decline aborts, accept writes, `--force` skips input, EOF declines.

**Non-blocking notes**
- Cross-provider prefixed slugs: the guard only consults the *resolved* provider's catalog. A value explicitly prefixed with a different known provider (e.g. resolved provider `openrouter` but a model id that isn't in OpenRouter's catalog as-written) will warn/confirm even though it's a legitimate explicit route. Since OpenRouter's catalog includes most vendor-prefixed ids this is mostly benign, but when the prefix names a known provider, checking that provider's catalog (in addition to the resolved one) would eliminate the spurious prompt.
- The decline path uses `sys.exit(1)`, consistent with the surrounding `set_config_value` error handling — but any non-CLI caller on a TTY would get a hard `SystemExit` instead of a signal/return value. Fine for the CLI; just noting the composition constraint if this function is ever called programmatically.
- `except Exception: return` in `_maybe_warn_unknown_model_slug` silently swallows catalog/import errors. Fail-open is the right default here, but a debug log would distinguish "unknown provider" (expected) from "catalog plumbing broke" (a real bug worth surfacing).

PR #97664, opened 2026-08-29
feat(approvals): bind transport requests to profile and conversation scope

aosoficial:
## Summary

Extends the existing host-owned approval transport request with canonical profile identity and an opaque conversation scope.

- adds `profile_name` and `conversation_scope_key` to `ApprovalRequest`
- derives scope inside the host from canonical profile + raw session key
- never exposes the raw session key to the selected transport or its pre/post lifecycle hooks
- binds profile and opaque scope into the request digest
- fails closed if profile resolution, normalization, validation, redaction, or request construction fails
- preserves `request.session_key` compatibility, but its value is now the opaque scope

This is a generic transport contract only. It does **not** add a broker, approval authority, deployment, task/run scope, or a way to bypass the built-in approval policy.

## Security properties

- Callers cannot override `conversation_scope_key`.
- Profile names use Hermes' canonical `normalize_profile_name()` and `validate_profile_name()` behavior.
- Empty, invalid, and reserved profile names deny before invoking a transport.
- Raw session values do not reach the transport request or selected transport hooks.
- Transport responses remain bound to the exact request ID and digest.
- Construction failures do not fall back to raw/builtin presentation.

## Scope distinction

This is deliberately **conversation scope**, not task/run scope. It does not overlap with the Kanban task/run scoping work in #87963.

## Verification

- focused approval transport suite: 41 passed
- lifecycle/plugin suites: 93 passed
- Ruff: passed
- `py_compile`: passed
- `git diff --check`: passed
- ty: no diagnostics added relative to the previous branch commit
- independent final review: APPROVE

A pre-existing logging-handler message can appear during one redaction teardown test; the same message reproduces on the prior commit and does not affect test results.


Enough1122:

**Overall:** Extends `ApprovalRequest` with a canonical `profile_name` and an opaque `conversation_scope_key` (SHA-256 of profile + raw session key). The raw session key is never exposed to transports or their pre/post lifecycle hooks; the scope binds profile + conversation into the request and its digest, and `_present_with_selected_transport` resolves the active profile and fails closed (deny, no transport call) on any resolution/validation error. Comprehensive tests.

**What it does**
- `make_approval_scope(profile_name, session_key)` → `sha256("hermes-approval-scope-v1\0<profile>\0<session>")`.
- `ApprovalRequest.create`: normalizes + validates `profile_name`, derives `conversation_scope_key`, stores the opaque scope into the compatibility `session_key` field, and adds `profile_name`/`conversation_scope_key`; `profile_name` defaults to `"default"`. Scope cannot be caller-supplied (unknown kwargs raise `TypeError`).
- `tools/approval.py`: resolves `get_active_profile_name()` inside the existing try/except (so errors → deny without invoking the transport) and passes the opaque `request.session_key` + `conversation_scope_key` to hooks instead of the raw session key.
- Tests: scope stable for same profile+session, differs across profile/session, digest changes accordingly, no raw-key leak (even in `repr`); scope override rejected; invalid/empty profiles rejected; canonical names accepted; legacy default; hooks receive 64-char opaque scope; profile-resolution error and empty profile both deny without transport; redaction-failure path unchanged; docs updated.

**Non-blocking notes**
- `ApprovalRequest.session_key` (compat) and `conversation_scope_key` now carry the identical value — two fields, one meaning. The compat naming is understandable for lifecycle/plugin consumers, but it's a footgun: a future reader may assume `session_key` is the real key. A deprecation comment on the field (or dropping the alias once consumers migrate) would prevent misuse.
- The scope hash is deterministic and unsalted: the same (profile, session) always yields the same scope, so a transport can correlate approval requests across time for one conversation. That determinism is likely what makes session-scoped approval continuity work — if so, an explicit comment on the tradeoff would help future security review.
- `ApprovalRequest.create` lazy-imports `hermes_cli.profiles` and validates inside the constructor. Fail-closed deny relies on callers wrapping `create` in try/except (as `_present_with_selected_transport` does). The `default` profile fallback means a direct caller that forgets to pass the active profile would silently scope to `"default"` — in-tree callers all pass the resolved profile, so this only matters for out-of-tree use.

PR #97665, opened 2026-08-29
fix(delegate): stop provider failures looking like iteration exhaustion

### LINKED ISSUE #97655: [Bug]: Failed subagents are mislabeled status=completed with exit_reason=max_iterations
### Bug Description

A subagent that fails instantly — e.g. killed by `HTTP 400: <model> is not a valid model ID` on its very first API call — is reported to the parent as `status=completed` with `exit_reason=max_iterations` and the rendered claim "TRUNCATED: hit max_iterations — work may be incomplete". Both labels are false: the child made 1 API call in 0.74 seconds, hit no iteration cap, and produced no work.

This is not cosmetic: the mislabel actively misleads any agent (or human) diagnosing the failure. In the incident that surfaced this, a session diagnosing repeated delegation failures had to see through "completed + hit max_iterations" to find the real cause (a config typo), and the wrong `exit_reason` steers recovery toward "the model is too slow / task too big" instead of "the model ID is invalid."

### Steps to Reproduce

1. Set `delegation.model` to a slug the provider rejects (e.g. OpenRouter `upstage/solar-pro-4`).
2. Dispatch a `delegate_task` batch with one task.
3. Read the batch-completion report.

Result from a real run on my install (`deleg_7cfa5756`, 2026-08-22):

```
```

And in the live transcript:

```
```

`api_calls=1` in `0.74s` with an HTTP 400 summary, labeled `completed` + `max_iterations`.

### Expected Behavior

A child whose provider call failed (non-empty `error`, `failed: True` in the child result, or a `model_not_found`-classified summary) should be reported as `status=failed` (or `error`), with `exit_reason` reflecting the actual termination (`error` / `provider_rejected`), not `max_iterations`. The `truncated` flag should be false — nothing was truncated; nothing ran.

### Actual Behavior

`status=completed`, `exit_reason=max_iterations`, plus the rendered warnings "TRUNCATED: hit max_iterations — work may be incomplete" and "[TRUNCATED — subagent hit its iteration cap …]" — all three factually wrong for this failure mode.

### Affected Component

- Tools (delegate_task — `_run_single_child` status derivation, `tools/delegate_tool.py`)

### Debug Report

### Operating System

macOS 26.6.2

### Python Version

3.11.15

### Hermes Version

Two compounding mis-derivations in `tools/delegate_tool.py`:

1. **Status** (`:3063-3066`): the status branch keys on the summary string alone — `elif summary and not _empty_sentinel: status = "completed"`. For a provider-rejected child, the child's conversation loop returns the error text as `final_response` (its terminal label), so the error string *is* the summary, and the child is labeled `completed`. The `result.get("failed")` / `result.get("error")` fields (`agent/conversation_loop.py:816-820` sets `failed: True` + `error:` on the structured terminal result) are never consulted in this branch.

2. **Exit reason** (`:3100-3104`): `exit_reason` is derived from the child's `completed` flag — `completed=False` is *assumed* to mean the iteration budget was exhausted (`else: exit_reason = "max_iterations"`). But `completed: False` is also what every failure path returns (`conversation_loop.py:816`, `:1461`, `:1506`), so any failure gets reported as budget exhaustion.

The renderer then trusts those fields (`tools/process_registry.py:3028`: `r_truncated = r.get("truncated") or r.get("exit_reason") == "max_iterations"`) and produces the "TRUNCATED: hit max_iterations" text.

Note the existing comment at `:3055-3058` (empty-response sentinel) already recognizes this class of problem for one special case: "(empty)" summaries are treated as failure. The same reasoning applies to error-shaped final responses — the fix is to check `result.get("failed")` / `result.get("error")` *before* trusting the summary-presence heuristic.

### Proposed Fix (optional)

In `_run_single_child`'s status derivation (`tools/delegate_tool.py:3063`), restructure the branch order to:

```python
if interrupted:
    status = "interrupted"
elif result.get("failed") or result.get("error"):
    status = "failed"          # provider rejections, terminal failures
elif summary and not _empty_sentinel:
    status = "completed"
else:
    status = "failed"
```

and derive `exit_reason` from the failure when present (`"error"` / `classified reason` if available) rather than defaulting to `max_iterations` whenever `completed` is False. This keeps the legitimate truncation case (budget exhausted *with* a summary) working — that path still has `failed` unset and a real summary — while provider-rejected children get an honest label. The batch renderer needs no change once the fields are correct, though a test asserting the 400 case renders `✗ ... status=failed` (not `⚠ ... TRUNCATED`) would pin the contract.

See live transcript excerpt in Bug Description; identical artifacts exist for `deleg_71376970`, `deleg_907207b2` (3 tasks), `deleg_5cab5c7f` — all single-API-call children reported `completed` + `max_iterations` after instant HTTP 400s.

fangliquanflq:
## What does this PR do?

Provider-rejected subagents are now reported as failed instead of completed and truncated by a nonexistent iteration limit. The delegate result classifier now gives structured terminal failures precedence over a non-empty error-text summary, preserves a machine-readable failure reason when available, and leaves explicit soft outcomes alone.

### Symptom

A child rejected by the provider on its first API call could be rendered as `status=completed`, `exit_reason=max_iterations`, and `TRUNCATED`, even though it produced no work and never exhausted its iteration budget.

### Impact

The parent agent and operator receive the wrong diagnosis and may retry with a larger budget instead of correcting the provider or model configuration.

### Bug Cause

**Trigger:** `tools/delegate_tool.py` / `_run_single_child` status and exit-reason derivation

**Causal chain:**
1. A provider failure returns a structured child result with `failed=True`, an `error`, `completed=False`, and the error text in `final_response`.
2. `_run_single_child` treats any non-empty `final_response` as usable output and labels the child completed without checking the structured failure fields.
3. The same code maps every `completed=False` result to `max_iterations`, causing the async renderer to add a false truncation warning.

**Why it is wrong:** The summary-presence heuristic and completed flag were allowed to override the child's authoritative terminal failure metadata.

**Working sibling / contrast:** Genuine budget exhaustion has no terminal failure flag and still follows the existing summarized-but-truncated path.

**Ruled out:** The async renderer is not inventing the failure classification; it renders the `status`, `exit_reason`, and `truncated` fields produced by `_run_single_child`.

### Fix

Classify explicit `failed=True` results, plus legacy error-only terminal results, before considering summary text. Use the child's machine-readable `failure_reason` when present, otherwise `error`, so provider failures no longer become `max_iterations`. An explicit `failed=False` still protects soft outcomes such as compression deferral. Structured-output delegations now skip schema retries for an already-terminal result and preserve authoritative failure metadata when the bounded schema retry itself fails.

## Related Issue

Closes #97655

## Type of Change

## Changes Made

- `tools/delegate_tool.py` - prioritize structured child failures when deriving delegation status, exit reason, and truncation.
- `tests/tools/test_delegate.py` - reproduce a one-call provider rejection and assert its honest failed classification.
- `tests/tools/test_delegate_output_schema.py` - cover terminal failures before and during the bounded structured-output retry.

## How to Test

1. Run the focused delegation test suite.
2. Confirm the provider-failure regression reports `status=failed`, `exit_reason=model_not_found`, and `truncated=false`.

```bash
scripts/run_tests.sh tests/tools/test_delegate_output_schema.py tests/tools/test_delegate.py tests/tools/test_async_delegation.py -q
```

Result: 119 passed.

### Code

- [x] I've updated relevant documentation - N/A; no user-facing documentation changed
- [x] I've considered cross-platform impact per the compatibility guide; the fix is platform-independent Python logic
- [x] I've updated tool descriptions/schemas if I changed tool behavior - N/A; the schema is unchanged

## Screenshots / Logs

N/A; the focused regression test covers the structured result consumed by the renderer.

### DIFF
--- a/tests/tools/test_delegate.py
+++ b/tests/tools/test_delegate.py
@@ -667,6 +667,39 @@ class TestDelegateObservability(unittest.TestCase):
             result = json.loads(delegate_task(goal="Test empty sentinel", parent_agent=parent))
             self.assertEqual(result["results"][0]["status"], "failed")
 
+    def test_provider_failure_is_not_reported_as_iteration_exhaustion(self):
+        """A structured provider failure must outrank its error-text summary."""
+        parent = _make_mock_parent(depth=0)
+
+        with patch("run_agent.AIAgent") as MockAgent:
+            mock_child = MagicMock()
+            mock_child.model = "upstage/solar-pro-4"
+            mock_child.session_prompt_tokens = 0
+            mock_child.session_completion_tokens = 0
+            mock_child.run_conversation.return_value = {
+                "final_response": "HTTP 400: model is not a valid model ID",
+                "completed": False,
+                "failed": True,
+                "error": "HTTP 400: model is not a valid model ID",
+                "failure_reason": "model_not_found",
+                "interrupted": False,
+                "api_calls": 1,
+                "messages": [],
+            }
+            MockAgent.return_value = mock_child
+
+            result = json.loads(
+                delegate_task(goal="Test rejected provider model", parent_agent=parent)
+            )
+            entry = result["results"][0]
+
+            self.assertEqual(entry["status"], "failed")
+            self.assertEqual(entry["exit_reason"], "model_not_found")
+            self.assertFalse(entry["truncated"])
+            self.assertEqual(
+                entry["error"], "HTTP 400: model is not a valid model ID"
+            )
+
 
 class TestSubagentCostRollup(unittest.TestCase):
     """Port of Kilo-Org/kilocode#9448 — parent's session_estimated_cost_usd
--- a/tests/tools/test_delegate_output_schema.py
+++ b/tests/tools/test_delegate_output_schema.py
@@ -313,8 +313,34 @@ class TestRunSingleChildSchemaValidation:
         assert entry["exit_reason"] == "model_not_found"
         assert entry["truncated"] is False
         assert entry["error"] == "HTTP 400: model rejected"
+        assert entry["summary"] == "HTTP 400: model rejected"
         assert entry["schema_valid"] is False
         assert entry["schema_retries"] == 1
+
+    def test_provider_failure_on_retry_clears_stale_summary(self):
+        child = _StubChild(
+            [
+                "not json at all",
+                {
+                    "completed": False,
+                    "failed": True,
+                    "error": "HTTP 400: model rejected",
+                    "failure_reason": "model_not_found",
+                    "interrupted": False,
+                    "api_calls": 1,
+                    "messages": [],
+                },
+            ]
+        )
+        child._delegate_output_schema = ADDRESS_SCHEMA
+
+        entry = _run(child)
+
+        assert entry["status"] == "failed"
+        assert entry["summary"] == ""
+        assert entry["exit_reason"] == "model_not_found"
+        assert entry["truncated"] is False
+        assert entry["error"] == "HTTP 400: model rejected"
 
 
 # ---------------------------------------------------------------------------
--- a/tests/tools/test_delegate_output_schema.py
+++ b/tests/tools/test_delegate_output_schema.py
@@ -169,9 +169,11 @@ class _StubChild:
 
     def run_conversation(self, user_message, task_id=None, **_kwargs):
         self.calls.append(user_message)
-        text = self.responses.pop(0)
+        response = self.responses.pop(0)
+        if isinstance(response, dict):
+            return response
         return {
-            "final_response": text,
+            "final_response": response,
             "completed": True,
             "api_calls": 1,
             "messages": [],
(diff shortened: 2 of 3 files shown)


Enough1122:

Reviewed `tools/delegate_tool.py` + tests for PR #97665. The direction is sound: a structured terminal failure (provider rejection) should outrank a non-empty error-text summary and not be mislabeled as `max_iterations` exhaustion. The `_is_terminal_child_result` helper and its precedence in `_run_single_child` look correct, and the new tests cover the first-call and retry-turn cases well.

A couple of non-blocking observations:

1. **Retry outcome merge does not propagate `final_response`.** In the retry branch, when `_is_terminal_child_result(_retry_result)` is true, the enumerated outcome keys (`completed`, `failed`, `error`, `failure_reason`, `interrupted`) are copied into `result`, but `final_response` (the user-facing text) is not. If a terminal failure surfaces on the retry turn, `result["final_response"]`/summary stays at the first turn's value, so the parent can see a stale non-error text while `error` holds the real message. Consider also copying `final_response` (or clearing it) for consistency with the terminal outcome.

2. **Merged keys can leave contradictory state.** When the first turn set `completed=True` (e.g. the string-response → dict conversion in the schema path) and the retry result is a terminal failure that does not carry `completed`, the merged result ends up with both `completed=True` and `failed=True`. The new precedence handles this inside `_run_single_child`, but any downstream code that reads `result["completed"]` directly would see a misleading value. Overriding/clearing `completed` on a terminal merge would be more robust.

Both are minor; the fix as written is correct for the reported symptom and well tested.

fangliquanflq:
Thanks for the careful review. I investigated both points against the retry path and its consumers.

1. Accepted for the empty terminal-response case. Non-empty retry text was already copied before the outcome merge, but an empty terminal `final_response` left the first turn's invalid text in `summary`. Commit `420d12bcc0` now treats a terminal retry as authoritative even when its response is empty, and adds coverage for both non-empty propagation and stale-summary clearing.
2. No additional change is needed for `completed`. The merged child result is local to `_run_single_child`; callers receive the normalized entry. Terminal precedence determines `status`, `exit_reason`, and `truncated`, so an omitted retry `completed` value cannot escape through the public delegation result as completion or iteration exhaustion. The focused tests cover those returned fields.

Validation: `scripts/run_tests.sh tests/tools/test_delegate_output_schema.py tests/tools/test_delegate.py tests/tools/test_async_delegation.py -q` — 120 passed. An independent committed-tip review also found no remaining issues.

PR #97667, opened 2026-08-29
fix(tools): surface config-level model_not_found notices in delegation batch reports

### LINKED ISSUE #97654: [Bug]: Delegation model rejections (model_not_found) are invisible to the user — no config-level notice
### Bug Description

When the configured delegation model is rejected by the provider (`HTTP 400: <model> is not a valid model ID`), every subagent dies within ~1 second, and the parent-facing batch report buries the cause inside a per-task block. No dedicated, actionable notice reaches the user or the agent — nothing says "your Subagent Model is invalid; check Settings → Advanced."

The failure repeats identically for every delegation attempt until a human notices the pattern in batch dumps. On my install this persisted for roughly a week (the user changed a setting, typo'd the slug, and delegation was silently dead across many sessions).

### Steps to Reproduce

1. Set `delegation.model` to a slug the provider will reject (e.g. on OpenRouter: `upstage/solar-pro-4` — note the hyphen; the valid slug is `upstage/solar-pro4`).
2. In any session, invoke `delegate_task` with a single task.
3. Observe the batch-completion report.

Actual result (real log from my install, `deleg_7cfa5756`):

```
Full live transcript (complete tool/assistant trace): ~/.hermes/profiles/coder/cache/delegation/live/deleg_7cfa5756/task-0.log
```

The only place the root cause appears is the `HTTP 400:` line inside the summary text. The header simultaneously mislabels the failure as `completed` and blames `max_iterations` (see companion issue on mislabeled statuses).

### Expected Behavior

When the delegation model itself is rejected (`model_not_found` classification — `agent/error_classifier.py:404` already classifies "is not a valid model" this way), the batch report should include a dedicated, top-level, actionable notice, something like:

```
⚠ SUBAGENT MODEL REJECTED: the configured Subagent Model "upstage/solar-pro-4"
  was rejected by provider "openrouter" (HTTP 400: not a valid model ID).
  Every task in this batch failed for this reason before doing any work.
  Check Settings → Advanced → Subagent Model (or: hermes config get delegation.model).
```

This notice should appear regardless of per-task status labels, since the root cause is config-level, not task-level.

### Actual Behavior

The 400 text appears only as the child's `summary` string inside the standard per-task template, with no detection, no escalation, and no config-level guidance. A user scanning the batch sees a wall of task blocks; the signal that *every* task failed identically before doing any work is easy to miss. In multi-task batches the 400 repeats once per task, amplifying noise while hiding the meaning.

### Affected Component

- Tools (delegate_task / async delegation batch reporting — `tools/process_registry.py` `render`/summary path, `tools/delegate_tool.py` result assembly)

### Debug Report

### Operating System

macOS 26.6.2

### Python Version

3.11.15

### Hermes Version

The batch report renderer (`tools/process_registry.py:3020-3063`, `:3080-3100`) treats each result entry independently and has no concept of "all tasks failed with the same provider rejection." The `model_not_found` classification exists in `agent/error_classifier.py:404-420` (it even documents that this classification should trigger failover), but nothing in the delegation batch path consumes the classification to produce a config-level notice. Meanwhile a default install has an empty provider-fallback chain, so nothing fails over either — the failure just repeats.

### Proposed Fix (optional)

In the batch aggregation/render path, when ≥1 task result carries a `model_not_found`-classified error naming the delegation model itself, prepend the config-level warning block above (single instance, not per-task). Optionally cross-reference the parent model's fallback chain: if it's empty (the default), say so — "no fallback chain configured, so no failover was attempted."

Live transcript for the example above (`deleg_7cfa5756/task-0.log`), 10 lines total — kickoff, start, then:

```
```

liuhao1024:
## What does this PR do?

When the configured `delegation.model` slug is rejected by the provider (e.g. a typo'd `upstage/solar-pro-4` where the valid slug is `upstage/solar-pro4`), every subagent in a `delegate_task` batch dies within ~1 second carrying the provider's rejection text as its summary, while the per-task blocks keep labelling the run `status=completed` + `TRUNCATED: hit max_iterations`. The config-level root cause stays buried inside the per-task summaries — nothing tells the user "your Subagent Model is invalid" (#97654).

This adds a single config-level notice to the batch completion report. A task counts as a rejection hit only when its summary/error text both matches a `model_not_found` pattern (reused from `agent.error_classifier._MODEL_NOT_FOUND_PATTERNS` — the same classification the failover path consumes, kept as the single source of truth) and names the configured delegation model id, so incidental prose about other models never triggers it. The notice is rendered once above the per-task blocks with the model id, the hit count, and the setting to fix.

## Related Issue

Fixes #97654

## Type of Change

## Changes Made

- `tools/process_registry.py` — added `_batch_model_rejection_notice()`: scans a batch's per-task summary/error text for `model_not_found` patterns naming the configured delegation model, returns the config-level notice lines (or `None`).
- `tools/process_registry.py` (`_format_async_delegation`, batch branch) — insert the notice after the `Role:/Model:` header line and before the per-task blocks, only when the batch has results.
- `tests/tools/test_async_delegation.py` — three regression tests: notice prepended before the task blocks (issue's exact scenario), no notice for clean summaries, and no notice when the pattern names a different model than the configured one.

## How to Test

1. `uv run --extra dev pytest tests/tools/test_async_delegation.py -q`
   Observed result: `29 passed` (26 pre-existing + 3 new).
2. Focused: `uv run --extra dev pytest tests/tools/test_async_delegation.py -q -k model_rejection`
   Observed result: `3 passed`.
3. `test_batch_model_rejection_notice_prepended` reproduces the issue's exact payload (two tasks, `status=completed`, `exit_reason=max_iterations`, summary `HTTP 400: upstage/solar-pro-4 is not a valid model ID`) and asserts the rendered block now contains `SUBAGENT MODEL REJECTED`, the model id, `2/2`, the `delegation.model` hint, and that the notice precedes `TASK 1/2`.
4. Lint: `ruff check tools/process_registry.py tests/tools/test_async_delegation.py` — all checks pass; `ty check tools/process_registry.py` shows no new diagnostics vs. the pre-change tree (57 = 57).

### Code

- [x] I've tested on my platform: macOS 26 (arm64)

- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (pure string rendering, no platform-specific code)

### DIFF
--- a/tests/tools/test_async_delegation.py
+++ b/tests/tools/test_async_delegation.py
@@ -881,3 +881,86 @@ def test_batch_truncation_banner_marks_only_truncated_task():
     # The header banner for task 2 appears after task 1's summary.
     assert banner_pos > clean_pos
 
+
+def test_batch_model_rejection_notice_prepended():
+    """A rejected delegation model must surface ONE config-level notice above
+    the per-task blocks instead of staying buried in each summary (#97654)."""
+    rejection = "HTTP 400: upstage/solar-pro-4 is not a valid model ID"
+    evt = _make_async_evt(
+        is_batch=True,
+        model="upstage/solar-pro-4",
+        goals=["task a", "task b"],
+        results=[
+            {
+                "task_index": 0,
+                "status": "completed",
+                "summary": rejection,
+                "api_calls": 1,
+                "duration_seconds": 0.74,
+                "exit_reason": "max_iterations",
+                "truncated": True,
+            },
+            {
+                "task_index": 1,
+                "status": "completed",
+                "summary": rejection,
+                "api_calls": 1,
+                "duration_seconds": 0.71,
+                "exit_reason": "max_iterations",
+                "truncated": True,
+            },
+        ],
+    )
+    text = format_process_notification(evt)
+    assert text is not None
+    assert "SUBAGENT MODEL REJECTED" in text
+    assert "upstage/solar-pro-4" in text
+    assert "2/2" in text
+    assert "delegation.model" in text
+    # The notice precedes the per-task blocks, not just trails them.
+    assert text.index("SUBAGENT MODEL REJECTED") < text.index("TASK 1/2")
+
+
+def test_batch_model_rejection_notice_absent_when_clean():
+    """Ordinary summaries must not grow a model-rejection notice."""
+    evt = _make_async_evt(
+        is_batch=True,
+        model="upstage/solar-pro4",
+        goals=["task a"],
+        results=[
+            {
+                "task_index": 0,
+                "status": "completed",
+                "summary": "did the work",
+                "api_calls": 3,
+                "exit_reason": "completed",
+                "truncated": False,
+            },
+        ],
+    )
+    text = format_process_notification(evt)
+    assert text is not None
+    assert "SUBAGENT MODEL REJECTED" not in text
+
+
+def test_batch_model_rejection_notice_requires_configured_model_in_text():
+    """A model_not_found pattern naming a DIFFERENT model than the configured
+    delegation model is task-level noise, not a config-level rejection."""
+    evt = _make_async_evt(
+        is_batch=True,
+        model="upstage/solar-pro4",
+        goals=["task a"],
+        results=[
+            {
+                "task_index": 0,
+                "status": "completed",
+                "summary": "HTTP 400: other/model-x is not a valid model ID",
+                "api_calls": 1,
+                "exit_reason": "max_iterations",
+                "truncated": True,
+            },
+        ],
+    )
+    text = format_process_notification(evt)
+    assert text is not None
+    assert "SUBAGENT MODEL REJECTED" not in text
--- a/tools/process_registry.py
+++ b/tools/process_registry.py
@@ -2960,6 +2960,48 @@ def _format_age(seconds: float) -> str:
     return f"{h}h" if m == 0 else f"{h}h{m}m"
 
 
+def _batch_model_rejection_notice(results, model):
+    """Config-level notice lines when a batch died on a rejected model id.
+
+    A typo'd ``delegation.model`` slug makes every subagent abort within a
+    second with the provider's rejection text as its summary, while the
+    per-task blocks keep labelling it ``status=completed`` + TRUNCATED — the
+    config-level cause stays buried in the batch dump (#97654). A task counts
+    as a rejection hit only when its summary/error text both names the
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed `tools/process_registry.py` + tests for PR #97667. The feature is well-motivated (config-level model rejection buried inside per-task blocks in batch notifications) and the heuristic is sensibly guarded: a task only counts as a rejection hit when the combined summary/error text both names the configured model and matches a `model_not_found` pattern, and the new tests cover the clean and false-positive (different model) cases.

A few non-blocking observations:

1. **Substring match on the model id is a loose equality test.** `model_l in text` will also match a model id that is a prefix of another (e.g. `.../pro4` matching `.../pro-4` text, or a short slug matching incidental prose). The `_MODEL_NOT_FOUND_PATTERNS` check reduces false positives, but a negative match can also silently miss a rejection if the provider wraps/escapes the model name so it no longer appears verbatim in the task text (case is handled; quoting/whitespace is not). Consider matching the model as a token with flexible separators (`/`, `-`, `_`, `:`) or documenting that this is a best-effort heuristic.

2. **The notice claims "failed for this reason before doing any work", but the hit test only inspects text.** Nothing verifies `api_calls == 1` or an empty working summary before the notice asserts the tasks "died before doing any work". Since `_MODEL_NOT_FOUND_PATTERNS` + the model-id match already strongly imply a pre-flight rejection, this is a wording/verification gap rather than a functional one — tightening the message ("…show a model_not_found rejection") would keep it accurate without extra checks.

3. **Private API coupling.** Importing `_MODEL_NOT_FOUND_PATTERNS` (a private module attribute) into `process_registry` couples the formatter to `error_classifier`'s internals. A shared public constant/classifier function (e.g. a `classify_error()` returning `model_not_found`) would be more maintainable than re-implementing pattern membership here. Non-blocking.

The placement of the notice above the per-task blocks and the clear `delegation.model` pointer are good UX; nothing blocking found.

teknium1:
Thanks @liuhao1024 — your PR was the first fix for #97654 (16h ahead of the version that merged in #99067). Your key improvement — importing _MODEL_NOT_FOUND_PATTERNS from agent/error_classifier instead of hand-copying the pattern list — was cherry-picked with your authorship preserved and merged in #99098 (merge commit b215e8d5f9), replacing the duplicated list. Your notice tests rode along, adapted to the config-reading renderer on main. Closing since the substance is landed under your name.

PR #97668, opened 2026-08-29
fix(agent): normalize relative MEDIA paths against the session cwd at finalize

YuhGuan:
## Problem

The system prompt asks for `MEDIA:/absolute/path`, but a model working inside a project directory routinely emits relative paths (`MEDIA: final_video.mp4`). Every consumer downstream assumes absolute:

- the messaging gateways' MEDIA regexes only match absolute paths → the file silently isn't delivered;
- the desktop resolves the path against the app's own cwd → a black, unplayable card (observed on a Windows deployment, reproduced identically on macOS).

## Fix

At turn finalize, rewrite **line-form** relative MEDIA paths by joining them onto the session's working directory — live terminal cwd first (`get_session_cwd`, tracks `cd` during the turn), then the configured session cwd. The matching assistant message in the histories is updated too, so the stored transcript stays consistent.

Three fences make this strictly no-worse-than-before:
1. **Line form only** (`MEDIA: <path>` alone on a line; quoted paths supported) — inline prose mentions untouched.
2. **Fenced code blocks untouched** — examples aren't deliveries.
3. **A rewrite only happens when the joined path exists on disk** — a wrong base-directory guess degrades to leaving the text unchanged, never to a broken rewrite.

Never raises: any surprise returns the original text.

## Tests

`tests/agent/test_media_path_normalizer.py` — 13 cases pinning the fences (relative/subdir/quoted-with-spaces rewrites; missing-file / absolute / home / drive / URL / fenced / inline / no-cwd all untouched; unclosed-fence state machine). Existing `test_turn_finalizer` suite passes unchanged. Verified end-to-end: a turn whose response contains a relative MEDIA line finalizes with the absolute path.

Companion desktop PR (resolves relative paths in already-stored transcripts): see cross-link comment.

### DIFF
--- a/agent/media_path_normalizer.py
+++ b/agent/media_path_normalizer.py
@@ -1,0 +1,137 @@
+"""Normalize relative ``MEDIA:`` paths in the final response.
+
+The system prompt asks the model to deliver files as ``MEDIA:/absolute/path``,
+but a model working inside a project directory routinely emits relative paths
+(``MEDIA: final_video.mp4``) instead. Every consumer downstream assumes
+absolute: the messaging gateways' MEDIA regexes only match absolute paths (the
+file silently isn't delivered), and the desktop renders a player that resolves
+the path against the app's own cwd — a black, unplayable card (observed on a
+Windows deployment; reproduced identically on macOS).
+
+This module rewrites such lines at turn-finalize time, joining the relative
+path onto the session's working directory. Three deliberate fences keep it
+strictly no-worse-than-before:
+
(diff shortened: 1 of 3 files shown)


YuhGuan:
Companion desktop PR (heals already-stored transcripts): #97669

alt-glitch:

Related: #97669 is the companion desktop change that resolves relative MEDIA paths in existing transcripts; this PR normalizes newly finalized paths for downstream delivery.

PR #97669, opened 2026-08-29
fix(desktop): resolve relative MEDIA paths against the owning session's cwd

YuhGuan:
## Problem

When the model emits a relative MEDIA path (`MEDIA: final_video.mp4` — routine while it works inside a project directory, despite the absolute-path instruction), the desktop renders a player card whose URL the main process resolves against the **app's own cwd**. The file is never found and the user sees a black, unplayable 0:00 card (`MediaError` code 4, `networkState` NO_SOURCE). Observed on a Windows deployment; reproduced identically on macOS.

## Fix

`Thread` already receives the session's live `cwd` — provide it through a new `MediaCwdContext` (per-thread, so split panes stay independent), and have `MediaAttachment` join relative paths onto it before resolving playback/open sources. `absolutizeMediaPath` guards itself: absolute, `~`, drive-letter and scheme'd forms (`data:`, `blob:`, `https:`, the media stream scheme) pass through untouched, as does everything when no cwd is available.

**This also heals transcripts that already contain relative MEDIA paths** — they start playing after the update without the agent resending anything.

## Tests

- `src/lib/media-absolutize.test.ts`: join for POSIX and pure-Windows cwds; absolute/home/drive/scheme/no-cwd pass-through.
- Renderer `tsc` clean; full desktop suite passes on a deployment carrying this exact change (4400+ tests).
- Live verification: a session whose stored message contains a relative MEDIA path rendered a dead card (readyState 0, error 4) before, and plays (readyState 4, correct duration) after.

Companion engine PR normalizes the paths at turn-finalize so messaging gateways (absolute-only delivery) benefit too: see cross-link comment.

### DIFF
--- a/apps/desktop/src/components/assistant-ui/markdown-text.tsx
+++ b/apps/desktop/src/components/assistant-ui/markdown-text.tsx
@@ -8,8 +8,9 @@ import {
   tailBoundedRemend
 } from '@assistant-ui/react-streamdown'
 import type { code as streamdownCode } from '@streamdown/code'
-import { type ComponentProps, memo, useEffect, useMemo, useState } from 'react'
+import { type ComponentProps, memo, useContext, useEffect, useMemo, useState } from 'react'
 
+import { MediaCwdContext } from '@/components/assistant-ui/media-cwd-context'
 import { ExpandableBlock } from '@/components/chat/expandable-block'
 import { PreviewAttachment } from '@/components/chat/preview-attachment'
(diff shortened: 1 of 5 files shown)


YuhGuan:
Companion engine PR (normalizes at finalize, benefits messaging gateways too): #97668

alt-glitch:

Related: #97668 normalizes newly finalized relative MEDIA lines for downstream delivery; this companion repairs resolution of relative paths already present in desktop transcripts.

PR #97670, opened 2026-08-29
fix(kanban): let block --kind consume a trailing reason

### LINKED ISSUE #88135: kanban: block --kind always fails with "unrecognized arguments" — typed block kind unreachable via CLI
## Summary

`hermes kanban block <id> --kind <kind> <reason>` always exits 2 with `unrecognized arguments: <reason>`, in every form tried. `block <id> <reason>` without `--kind` works. So the typed block kinds (`needs_input` / `capability` / `dependency` / `transient`) — the ones a human-review gate wants — are unreachable from the CLI.

## Repro matrix (Hermes, librarian profile, macOS 26.3.1, 2026-08-17)

| command | result |
|---|---|
| `hermes kanban block t_X --kind needs_input "multi word reason"` | `hermes: error: unrecognized arguments: multi word reason` (exit 2) |
| `hermes kanban block t_X --kind needs_input -- "reason"` | same error |
| `hermes kanban block t_X --kind needs_input confirm` | same error (single word too) |
| `hermes kanban block t_X confirm` (no `--kind`) | `Blocked t_X: confirm` — works |

`unblock --reason "..."` has no such issue.

## Root-cause hint

`hermes_cli/kanban.py` block parser: positional `reason` is `nargs="*"` and `--kind` is a normal option — argparse fails to consume the trailing positional once the option is present.

## Impact

Blocking still works (reason lands as a `BLOCKED:` comment), so this is cosmetic-to-moderate — but the kind classification that routes a block to a human (`needs_input`) vs. automated recovery (`dependency`) cannot be set from the CLI at all.

usefulish:
## Summary

`hermes kanban block <id> --kind <kind> <reason>` always failed with `unrecognized arguments: <reason>` (exit 2), in every form tried (`--kind <val> <reason>`, `--kind=<val> <reason>`, single-word reason, with `--`). `block <id> <reason>` without `--kind` works. Result: every typed block kind (`needs_input` / `capability` / `dependency` / `transient`) was unreachable from the CLI. Fixes #88135.

## Root cause

The `block` subparser declares the `reason` positional as `nargs='*'` plus a `--kind` option. argparse cannot consume tokens that follow an option once positional consumption has started, so the trailing reason was rejected as an extra argument.

## Fix

Hoist a `--kind <val>` / `--kind=<val>` token to the head of the token list (`_hoist_kind_token`) before parsing, so the `nargs='*'` reason positional consumes everything after it. Applied via a `_KanbanParser` argparse subclass (`parser_class=` on the kanban subparsers action) so every surface that builds through `build_parser()` gets it: the `hermes kanban` CLI, the gateway `/kanban` slash command, and `run_slash`.

Details:
- No-op for subcommands that don't use `--kind`.
- Stops at `--` (everything after it is literal).
- Leaves a `--kind` missing its value untouched so argparse reports the real error.
- The no-kind form (`block <id> <reason>`) parses unchanged (`kind=None`).
- `unblock --reason` was never affected.

## Tests

`tests/hermes_cli/test_kanban_cli.py`:
- `test_block_kind_parses_with_trailing_reason` — parse-level regression: `--kind <val>` + single/multi-word reason, `--kind=<val>` equals form, kind-only, and the no-kind backward-compat form.
- `test_block_kind_end_to_end_sets_kind_on_task` — `kanban_command` with `--kind needs_input` persists `block_kind='needs_input'` on the task.

## Verification

- `scripts/run_tests.sh tests/hermes_cli/test_kanban_cli.py -q` → 6/6 pass.
- Broader kanban set (48 files: all `tests/hermes_cli/test_kanban_*.py`, `test_subparser_routing_fallback.py`, `tests/tools/test_delegate_kanban_isolation.py`) → 301 passed, 0 failed, 2 skipped.
- `ruff check` clean; the two touched files carry pre-existing `ruff format` drift also present on `main` (not touched to keep the diff minimal).
- Live CLI E2E on a temp `HERMES_HOME`: `hermes kanban block t_X --kind needs_input "human input needed"` → exit 0, task `status=blocked`, `block_kind=needs_input`; no-kind form → exit 0, `block_kind` stays NULL.

Environment: macOS 26.3.1, Python 3.11 venv, hermes-agent 0.20.5 editable install.

### DIFF
--- a/hermes_cli/kanban.py
+++ b/hermes_cli/kanban.py
@@ -213,6 +213,47 @@ def _check_dispatcher_presence(
 # Argparse builder
 # ---------------------------------------------------------------------------
 
+def _hoist_kind_token(argv: list[str]) -> list[str]:
+    """Move a ``--kind <val>`` / ``--kind=<val>`` token to the front of the
+    token list so argparse can still consume a trailing ``nargs='*'`` reason.
+
+    argparse's ``nargs='*'`` positional cannot consume tokens that follow an
+    option once positional consumption has started: ``block t_X --kind
+    needs_input some reason`` fails with ``unrecognized arguments: some
+    reason`` because ``--kind`` sits mid-stream after the positional has
+    begun consuming. Hoisting the option to the head of the token list —
+    before every positional — lets the reason positional take everything
+    that follows. Stops at ``--`` (everything after it is literal) and never
+    rewrites a ``--kind`` that is missing its value (argparse reports the
+    real error instead). Only ``block`` uses ``--kind`` in the kanban tree.
+    """
+    argv = list(argv)
+    for i, tok in enumerate(argv):
+        if tok == "--":
+            break
+        if tok == "--kind":
+            if i + 1 < len(argv) and not argv[i + 1].startswith("-"):
+                return [tok, argv[i + 1]] + argv[:i] + argv[i + 2:]
+            break
+        if tok.startswith("--kind="):
+            return [tok] + argv[:i] + argv[i + 1:]
+    return argv
+
+
+class _KanbanParser(argparse.ArgumentParser):
+    """Argparse subclass that hoists ``--kind`` ahead of positionals.
+
+    Used as ``parser_class`` for the kanban subcommand tree so every
+    subparser (including ``block``) parses with the hoist applied. The
+    hoist is a no-op for every subcommand that does not use ``--kind``.
+    """
+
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: #88216 is an earlier open fix for #88135 using argument declaration ordering; this PR uses parse-time token hoisting. Both restore `block <id> --kind <kind> <reason>`; maintainer choice is needed.

PR #97064, opened 2026-08-28
fix(compression): stop the summary when the commit fence is cancelled (#96953)

JoaoMarcos44:
## Summary

A cancelled commit fence could not reach the **streaming summary call**. The fence is consulted once *before* summary dispatch (#76354 F6) and never again while the stream runs, and `CompressionCommitFence.try_cancel_before_commit()` — the only cancel path an async host can take without blocking its event loop — has no `cancel_event` parameter to signal with. The sole cancel source wired into `aux_interrupt_protection` was `agent._hard_interrupt_requested`, which no host sets when it merely stops waiting.

So when the gateway session-hygiene host gave up at its 600s total ceiling, the compression worker kept streaming a summary that the fence was **already guaranteed to refuse**:

- minutes of extra paid provider tokens per doomed attempt,
- one of the four bounded compression-pool slots (`_COMPRESS_EXECUTOR_MAX_WORKERS`) held for the whole overrun, so later legitimate compressions hit `pool_saturated`,
- a `commit_fence_cancelled` abort at the end with nothing committed.

In [#96953](https://github.com/NousResearch/hermes-agent/issues/96953) three of these workers overlapped, running **723s / 936s / 870s against a 600s ceiling** — every one of them doing its full summary *after* the host had already walked away.

This PR installs a `SummaryCancelSource` for the summary call: an Event-shaped OR of the hard-interrupt event and the fence's cancellation state. The protected provider seam (`_run_protected_sync_provider_call`) already polls its captured cancel source every 20ms, so the worker now unwinds within one poll of the host giving up — or of a `/stop`, `/restart`, or admission revoke — releasing its durable lease and pool slot and rolling the transcript back exactly as before.

The resulting `AuxiliaryExplicitCancellation` is classified by source: an operator stop stays `explicit_interrupt`, a fence cancellation is reported as `commit_fence_cancelled`.

## Relationship to #96967

Complementary, no overlap — different file, different layer.

| | #96967 (`gateway/run.py`) | this PR (`agent/conversation_compression.py`) |
|---|---|---|
| Layer | hygiene **retry** policy | compression **worker** cancellation |
| Fixes | re-arming after a cancel, wait extension, sibling attempts | the abandoned worker that keeps streaming |
| Issue bullets | "should not be immediately re-attempted", "cap consecutive attempts" | "should not stay demoted behind attempts that have already been aborted" |

[...]

PR #97671, opened 2026-08-29
fix(compression): stop superseded lean attempts from multiplying provider calls

### LINKED ISSUE #97648: [Bug]: Lean compression lacks an attempt call budget; oversized session reached 877 Luna calls
## Bug Description

Lean context compression treats one logical compression attempt as a multi-request batch but does not impose an attempt-scoped provider-call budget across that batch.

In the current implementation, one lean attempt performs:

1. one main summary LLM call; then
2. up to 28 sequential chunk-digest LLM calls.

For a sufficiently large compacted region, that is **up to 29 auxiliary model calls for one logical compression attempt**. If the host reaches its 600-second ceiling, loses ownership, or starts a stall-fallback/new attempt against the unchanged session, work from earlier attempts can continue consuming auxiliary calls even though its final commit will be rejected.

On one oversized real session, the pre-repair diagnostic snapshot showed repeated 600-second no-progress outcomes, stale attempt ownership, reclaimed compression locks, no usable compressed summary in the current pathological state, and a session-level usage ledger that had reached **877 recorded Luna compression calls**.

This report is specifically about the missing resource invariant:

> One logical compression attempt, including its summary, digest fan-out, fallback, cancellation, and retry behavior, must have a bounded model-call/token budget and must stop issuing provider calls after it loses ownership.

It is related to, but distinct from:

- #97180 — bounded concurrency for sequential lean digests;
- #97488 — ceiling timeout, detached workers, and false auto-reset;
- #96775 — durable backoff after explicit interruption;
- #45535 — interrupted-turn compression-state pollution.

Those reports do not include the 29-call attempt plan, the 877-call usage observation, or the same-session lean-versus-legacy A/B below.

## Steps to Reproduce

A content-redacted reproduction can use synthetic tool-heavy messages; no private transcript content is required.

1. Run Hermes Desktop with automatic compression enabled and lean tail mode:

   ```yaml
   compression:
     enabled: true
     tail_mode: lean
   ```

2. Route the auxiliary compression task to a working model/provider.
3. Grow a session until the active compacted region is large enough to hit `_LEAN_DIGEST_MAX_CHUNKS`.

   The observed pre-repair snapshot had:

   - 1,373 active messages;
   - 2,380,371 serialized digest-input characters;
   - no usable compressed summary committed for the pathological state.

4. Trigger automatic preflight compression.
5. Allow the host to reach the 600-second total ceiling, then resume/continue the unchanged session so another automatic or stall-fallback attempt starts.
6. Repeat turns/resumes while the oversized state remains authoritative.
7. Inspect:

   - `agent.log` / `errors.log` for no-progress, stale-attempt, commit-fence, and lock-reclaim events;
   - `session_model_usage` for `task='compression'` call and token totals;
   - active-message and compressed-summary state.

### Deterministic call-plan reproduction

At the installed revision:

```text
_LEAN_DIGEST_CHUNK_CHARS = 72,000
_LEAN_DIGEST_MAX_CHUNKS = 28
serialized digest input = 2,380,371 chars
ceil(2,380,371 / 72,000) = 34 chunks uncapped
capped digest calls = 28
main summary calls = 1
maximum model calls per lean attempt = 29
```

The calculation was executed against the imported live constants and asserted:

```json
{
  "serialized_chars": 2380371,
  "digest_chunk_chars": 72000,
  "uncapped_digest_chunks": 34,
  "capped_digest_calls": 28,
  "summary_calls": 1,
  "max_model_calls_per_lean_attempt": 29
}
```

### Controlled mitigation / A-B comparison

Change only the tail strategy while leaving compression enabled:

```yaml
compression:
  enabled: true
  tail_mode: legacy
```

Then retry compression on the same pathological session.

The controlled legacy run:

- started with 1,371 active messages and approximately 1,248,264 estimated tokens;
- made exactly **one** Luna compression request;
- completed in **166,375 ms**;
- committed successfully (`commit_status="committed"`);
- completed the in-place split (`split_status="in_place_committed"`);
- reduced the active context to 73 messages and approximately 90,885 rough tokens;
- left zero compression locks;
- allowed the session to resume normally.

## Expected Behavior

- A logical compression attempt has an explicit total auxiliary-call and token budget covering the summary, all digests, fallback-chain work, and retries.
- Every digest call checks attempt ownership/cancellation before contacting the provider.
- Once a host timeout, interrupt, or newer generation revokes an attempt, the old attempt issues no further auxiliary calls.
- A new attempt against the same source fingerprint does not start until the prior provider work has stopped, or the prior work is deterministically deduplicated/reused.
- The compressor either commits bounded progress or fails clearly with durable backoff/strategy change.
- Compression locks and Desktop compression state clear when the attempt is no longer active.
- Usage telemetry makes planned, completed, cancelled, and discarded calls visible per logical attempt.

## Actual Behavior

The affected session's cumulative usage ledger recorded the following compression row at the pre-repair snapshot:

```text
model: gpt-5.6-luna
task: compression
api_call_count: 877
input_tokens: 20,292,618
output_tokens: 2,017,868
```

This row is cumulative for that session's compression history; it is not a claim that all 877 calls belonged to one attempt. The failure signal is that this total coexisted with no usable compressed summary in the current pathological state, an unchanged oversized active context, and repeated new attempts that could each plan up to 29 more provider calls.

The same incident window contained:

- **34** `Context compression made no progress ... total wait 600.0s` warnings;
- **18** stale attempt-state restore warnings;
- **3** stale compression-lock reclamations;
- repeated aborted commit telemetry (`failure_class="commit_fence_cancelled"`);
- attempt telemetry whose total durations exceeded 1.1–1.2 million ms because detached/superseded work remained alive;
- unchanged authoritative session state, causing later attempts to repeat expensive work.

Sanitized representative logs:

```text
Context compression made no progress for 0.0s
(total wait 600.0s, ceiling 600.0s); continuing without compression

Skipping stale compressor attempt-state restore:
attempt generation N no longer owns the compressor (current: N+1).
A newer (stall-fallback) attempt's state is preserved.

Compression commit cancelled before session mutation.

Reclaimed stale compression lock for session=[REDACTED]
```

The commit fence prevented stale workers from mutating the session, which is correct for state integrity, but it did not prevent the discarded attempts from continuing to consume auxiliary calls before reaching that fence.

The successful legacy comparison was recorded in a separate compression usage row:

```text
api_call_count: 1
input_tokens: 92,927
output_tokens: 7,636
```

This is not merely a slow-compression symptom: it is provider-work amplification without committed progress.

## Affected Component

- Desktop

N/A — reproduced in Hermes Desktop.

## Debug Report

Private, force-redacted Nous diagnostics bundle (auto-deletes after 14 days):

https://support.nousresearch.com/diagnostics/7afc9a4f-d3aa-4543-91aa-6a5c251f53a7

No public paste was created because the logs contain unrelated private conversation content and local file paths.

## Operating System

Windows 11 (`10.0.26200`)

## Python Version

Python 3.11.15

## Hermes Version

Hermes Agent v0.20.6 (2026.8.27).

- Installed git HEAD: `ac6c8028e00d01ee2f299ba7fd03329c7f10382d`
- `hermes --version` reports the installation as up to date; reported upstream at submission time: `ee742fe1`

Successful same-session legacy telemetry:

```json
{
  "chunk_count": 0,
  "chunking": false,
  "commit_ms": 202,
  "commit_status": "committed",
  "event": "compression_attempt",
  "fallback_used": false,
  "split_status": "in_place_committed",
  "total_duration_ms": 166375
}
```

Successful context reduction:

```text
context compression done:
messages=1371->73
rough_tokens=~90,885
awaiting_real_usage=true
```

No transcript content, credentials, account identifiers, private project names, or session identifiers are included in this report.

### 1. Lean mode multiplies one logical attempt into up to 29 provider calls

`agent/context_compressor.py:76-80` documents that `_generate_summary` is one call per compression run and lean mode additionally calls `_build_chunk_digests`.

`agent/context_compressor.py:1092-1094` sets:

```python
_LEAN_DIGEST_CHUNK_CHARS = 72_000
_LEAN_DIGEST_MAX_CHUNKS = 28
_LEAN_DIGEST_MAX_TOKENS = 1_400
```

`ContextCompressor._build_chunk_digests` at `agent/context_compressor.py:4737-4793` calculates up to 28 chunks and loops over them sequentially. Each iteration directly calls `call_llm(..., task="compression")` at lines 4768-4776.

`_augment_summary_lean` at lines 4795-4813 invokes those digest calls after the main summary already exists.

There is no attempt-scoped call counter or token reservation spanning the summary plus all digest calls.

### 2. Digest failures degrade and continue rather than stopping the batch

`_build_chunk_digests` catches each exception independently at lines 4784-4787, emits a placeholder, and continues the loop. That is useful for summary availability, but it also means a degraded attempt can continue contacting the provider for every remaining chunk unless a separate cancellation check stops it.

### 3. Host timeout can release the lock and start fallback before the old worker exits

`agent/conversation_compression.py:1428-1453` releases the cancelled worker's durable compression lock and can immediately run a fallback attempt.

Lines 1471-1473 explicitly leave the old future on the shared pool because its commit is fenced:

```python
# Leave the future on the shared pool: fence cancel won, so a late
# commit cannot land ...
return messages, _resolve_fallback_prompt()
```

The fence protects the database commit boundary, but it does not itself cancel future digest-provider calls inside the detached worker.

### 4. Ownership is checked when restoring/committing state, not before every provider call

The attempt-generation guard in `agent/conversation_compression.py:362-381` prevents stale attempts from overwriting newer compressor attributes, and commit fences prevent stale mutation. Those are state-integrity protections, not resource-consumption protections.

The observed result is therefore internally consistent with the code:

1. one attempt can plan 29 model calls;
2. the host times out or supersedes it;
3. the old worker cannot commit but can remain alive;
4. a new attempt starts against unchanged state;
5. provider calls multiply while committed summaries remain unchanged.

## Proposed Fix

1. Add an attempt-scoped compression budget object covering:
   - summary calls;
   - digest calls;
   - fallback-chain calls;
   - total input/output token reservations;
   - wall-clock and cancellation state.
2. Check attempt ownership and cancellation immediately before every `call_llm` in `_build_chunk_digests`.
3. Propagate host cancellation into the auxiliary transport instead of relying only on a later commit fence.
4. Do not release the durable lock/start a same-source retry until the prior worker has stopped issuing provider calls, unless requests are deduplicated by a stable source fingerprint.
5. Persist a bounded backoff/strategy decision after ceiling failure so the unchanged session cannot immediately repeat the same 29-call plan.
6. Track per-attempt `planned_calls`, `started_calls`, `completed_calls`, `cancelled_calls`, `discarded_calls`, and token totals in compression telemetry.
7. Consider automatically falling back to the one-call legacy strategy when the lean plan exceeds a configurable call/token budget. Bounded concurrency from #97180 can reduce latency, but concurrency alone does not bound total provider work.

## Willingness to Submit a PR

I can help validate a maintainer fix against the redacted reproduction and the same-session lean-versus-legacy acceptance criteria.

fangliquanflq:
## What does this PR do?

Lean compression can currently issue up to 29 auxiliary model requests in one logical attempt, and a worker that loses its commit fence can continue issuing chunk-digest requests after its result becomes unpublishable. This change gives each attempt one context-local eight-call budget, coarsens digest segmentation to cover the whole source within the remaining slots, and propagates commit-fence cancellation into both queued and in-flight auxiliary work.

### Symptom

An oversized lean-mode session can repeatedly time out without committing progress while detached attempts continue consuming compression requests.

### Impact

Affected sessions can accumulate substantial auxiliary-model usage and latency without producing a usable compressed summary. The reported session recorded 877 cumulative compression calls while repeated attempts could each plan up to 29 more calls.

### Bug Cause

**Trigger:** `agent/context_compressor.py:_build_chunk_digests` after a large lean-mode summary, combined with host timeout cancellation in `agent/conversation_compression.py:compress_context`.

**Causal chain:**
1. Lean mode issues one summary request and then plans as many as 28 sequential digest requests.
2. The host can cancel the commit fence and detach the worker after the progress ceiling, but the digest loop does not consult that fence before another provider request.
3. A fallback or later attempt starts against unchanged state while the old worker continues consuming auxiliary requests whose commit will be rejected.

**Why it is wrong:** The commit fence protects session mutation but did not bound or revoke provider work, so resource consumption outlived attempt ownership.

**Working sibling / contrast:** Legacy mode uses one summary request and therefore does not have the lean digest fan-out. The same reported session completed with one legacy compression request.

**Ruled out:** The 877-call figure is not treated as one attempt. The deterministic live constants still show the per-attempt amplification: one summary plus up to 28 digest requests.

### Fix

Each built-in compression call now establishes an attempt-local provider budget using a `ContextVar`, so overlapping detached and fallback workers cannot share counters. Summary retries and digest calls reserve from the same eight-call cap. Digest chunks are made coarser when fewer slots remain, preserving coverage of the full compacted region instead of dropping its newest tail. Commit-fence cancellation is also passed to the auxiliary transport and checked before each digest dispatch, stopping both in-flight and subsequent provider work once ownership is lost. Plugin context engines do not receive the private cancellation argument unless they explicitly opt into that signature.

## Related Issue

Closes #97648

## Type of Change

## Changes Made

- `agent/context_compressor.py` - add the attempt-local call budget, budget-aware digest coarsening, and pre-dispatch cancellation checks.
- `agent/conversation_compression.py` - propagate commit-fence cancellation through the built-in compressor and auxiliary request scope without changing plugin kwargs.
- `tests/agent/test_compression_attempt_ownership.py` - cover total digest admission, full-region coarsening, cancellation, and plugin signature isolation.

## How to Test

1. Run the deterministic attempt-budget and ownership tests.
2. Run the compressor, concurrent timeout/fence, and worker-isolation suites.

```bash
scripts/run_tests.sh tests/agent/test_context_compressor.py tests/agent/test_compression_concurrent_fork.py tests/agent/test_compression_worker_isolation_76354.py tests/agent/test_compression_attempt_ownership.py -q
```

Result: 211 tests passed before the final focused test addition; the updated focused file then passed all 14 tests.

### Code

- [x] I've run the relevant repository test entry and all selected tests pass

- [x] Relevant documentation update is N/A; behavior and internal contract are documented in code and tests
- [x] `CONTRIBUTING.md` and `AGENTS.md` updates are N/A; no contributor workflow changed
- [x] I've considered cross-platform impact; the implementation uses Python `ContextVar` and existing cancellation primitives
- [x] Tool description/schema updates are N/A; no model tool changed

## Screenshots / Logs

N/A - verified with deterministic automated tests.

### DIFF
--- a/agent/context_compressor.py
+++ b/agent/context_compressor.py
@@ -4813,9 +4813,9 @@ Summary generation was unavailable, so this is a best-effort deterministic fallb
         Splits the region into ``_LEAN_DIGEST_CHUNK_CHARS`` chunks (capped at
         ``_LEAN_DIGEST_MAX_CHUNKS`` — beyond that, earliest chunks are merged
         coarser) and digests each with the compression LLM. Any chunk failure
-        degrades to a placeholder naming the message range; the whole call
-        never raises. Chunks run sequentially on the same transport as the
-        main summary.
+        degrades to a placeholder naming the message range. Explicit attempt
+        cancellation still propagates to the host. Chunks run sequentially on
+        the same transport as the main summary.
         """
         text = _serialize_turns_for_digest(
             turns, getattr(self, "_lean_pristine_tools", None),
@@ -4842,7 +4842,7 @@ Summary generation was unavailable, so this is a best-effort deterministic fallb
             if not _start_compression_provider_call():
                 logger.info(
                     "Stopping lean chunk digests after %d/%d segment(s): "
-                    "compression attempt provider budget cancelled or exhausted",
+                    "compression attempt provider budget exhausted",
                     ci,
                     n_chunks,
                 )
--- a/agent/context_compressor.py
+++ b/agent/context_compressor.py
@@ -1102,7 +1102,6 @@ class _CompressionAttemptProviderBudget:
     def __init__(self, max_calls: int, cancel_check=None) -> None:
         self.max_calls = max(1, int(max_calls))
         self.started_calls = 0
-        self.completed_calls = 0
         self._cancel_check = cancel_check
 
     @property
@@ -1124,10 +1123,6 @@ class _CompressionAttemptProviderBudget:
             return False
         self.started_calls += 1
         return True
-
-    def finish_call(self) -> None:
-        self.completed_calls += 1
-
 
 _COMPRESSION_ATTEMPT_PROVIDER_BUDGET: contextvars.ContextVar[
     Optional[_CompressionAttemptProviderBudget]
@@ -1154,14 +1149,9 @@ class _CompressionAttemptBudgetExhausted(RuntimeError):
 
 
 def _start_compression_provider_call() -> bool:
+    """Admit a call, preserving legacy unbounded behavior outside ``compress``."""
     budget = _COMPRESSION_ATTEMPT_PROVIDER_BUDGET.get()
     return budget is None or budget.try_start_call()
-
-
-def _finish_compression_provider_call() -> None:
-    budget = _COMPRESSION_ATTEMPT_PROVIDER_BUDGET.get()
-    if budget is not None:
-        budget.finish_call()
 
 _LEAN_DIGEST_PROMPT = """You are writing one segment of a detailed session log for an AI agent's context checkpoint. Digest the transcript segment below.
 
@@ -4867,8 +4857,6 @@ Summary generation was unavailable, so this is a best-effort deterministic fallb
             except Exception as exc:
                 logger.warning("lean chunk digest %d/%d failed: %s", ci + 1, n_chunks, exc)
                 body = f"[digest unavailable for segment {ci + 1}/{n_chunks} — recover via session_search]"
-            finally:
-                _finish_compression_provider_call()
             digests.append(f"### Segment {ci + 1}/{n_chunks}\n{body}")
         if not digests:
             return ""
@@ -5333,7 +5321,6 @@ This compaction should PRIORITISE preserving all information related to the focu
                 with aux_interrupt_protection():
                     response = call_llm(**call_kwargs)
             finally:
-                _finish_compression_provider_call()
                 route_known = bool(_aux_route.get("provider") and _aux_route.get("model"))
                 _aux_provider = _aux_route.get("provider") or self.provider or ""
                 _aux_model = _aux_route.get("model") or self.summary_model or self.model or ""
--- a/agent/context_compressor.py
+++ b/agent/context_compressor.py
@@ -1148,10 +1148,24 @@ class _CompressionAttemptBudgetExhausted(RuntimeError):
     pass
 
 
+def _raise_if_compression_attempt_cancelled() -> None:
+    budget = _COMPRESSION_ATTEMPT_PROVIDER_BUDGET.get()
+    if budget is not None and budget.is_cancelled:
+        raise AuxiliaryExplicitCancellation()
+
+
 def _start_compression_provider_call() -> bool:
     """Admit a call, preserving legacy unbounded behavior outside ``compress``."""
     budget = _COMPRESSION_ATTEMPT_PROVIDER_BUDGET.get()
-    return budget is None or budget.try_start_call()
+    if budget is None:
+        return True
+    _raise_if_compression_attempt_cancelled()
+    admitted = budget.try_start_call()
+    if not admitted:
+        # ``try_start_call`` also checks cancellation. Re-check so a cancel
+        # racing admission is never mistaken for ordinary budget exhaustion.
+        _raise_if_compression_attempt_cancelled()
+    return admitted
 
 _LEAN_DIGEST_PROMPT = """You are writing one segment of a detailed session log for an AI agent's context checkpoint. Digest the transcript segment below.
 
@@ -4813,8 +4827,9 @@ Summary generation was unavailable, so this is a best-effort deterministic fallb
         budget = _COMPRESSION_ATTEMPT_PROVIDER_BUDGET.get()
         digest_call_limit = _LEAN_DIGEST_MAX_CHUNKS
         if budget is not None:
+            _raise_if_compression_attempt_cancelled()
             digest_call_limit = min(digest_call_limit, budget.remaining_calls)
-            if digest_call_limit <= 0 or budget.is_cancelled:
+            if digest_call_limit <= 0:
                 return ""
         if n_chunks > digest_call_limit:
             chunk_size = (len(text) + digest_call_limit - 1) // digest_call_limit
@@ -4858,6 +4873,7 @@ Summary generation was unavailable, so this is a best-effort deterministic fallb
                 logger.warning("lean chunk digest %d/%d failed: %s", ci + 1, n_chunks, exc)
                 body = f"[digest unavailable for segment {ci + 1}/{n_chunks} — recover via session_search]"
             digests.append(f"### Segment {ci + 1}/{n_chunks}\n{body}")
+        _raise_if_compression_attempt_cancelled()
         if not digests:
             return ""
         return (
@@ -5391,6 +5407,7 @@ This compaction should PRIORITISE preserving all information related to the focu
             summary = _reinject_pruned_skill_markers(summary, _pruned_skill_names)
             summary = self._ground_historical_task_snapshot(summary, turns_to_summarize)
             summary = self._augment_summary_lean(summary, turns_to_summarize)
+            _raise_if_compression_attempt_cancelled()
             self._validate_summary_user_provenance(summary, has_user_turn)
             # Store for iterative updates on next compaction
             self._previous_summary = summary
@@ -7588,13 +7605,15 @@ This compaction should PRIORITISE preserving all information related to the focu
     ) -> List[Dict[str, Any]]:
         """Run one compression attempt under a total provider-call budget."""
         with _compression_attempt_provider_budget(cancel_check=_cancel_check):
-            return self._compress_impl(
+            result = self._compress_impl(
                 messages,
                 current_tokens=current_tokens,
                 focus_topic=focus_topic,
                 force=force,
                 memory_context=memory_context,
             )
+            _raise_if_compression_attempt_cancelled()
+            return result
 
     def _compress_impl(
         self,
--- a/agent/context_compressor.py
+++ b/agent/context_compressor.py
@@ -1092,7 +1092,76 @@ def _build_recovery_footer(session_id: str, region_len: int) -> str:
 _LEAN_DIGEST_CHUNK_CHARS = 72_000      # ~18K tokens of region per chunk
 _LEAN_DIGEST_MAX_CHUNKS = 28
 _LEAN_DIGEST_MAX_TOKENS = 1_400        # per-chunk digest cap (~13:1 ratio)
+_LEAN_ATTEMPT_MAX_PROVIDER_CALLS = 8   # summary/fallback + all chunk digests
 _LEAN_DIGESTS_HEADING = "## Detailed Session Log (chunked digests, oldest first)"
+
+
+class _CompressionAttemptProviderBudget:
+    """Attempt-local provider-call admission shared by summary and digests."""
+
+    def __init__(self, max_calls: int, cancel_check=None) -> None:
+        self.max_calls = max(1, int(max_calls))
+        self.started_calls = 0
+        self.completed_calls = 0
+        self._cancel_check = cancel_check
+
+    @property
+    def remaining_calls(self) -> int:
+        return max(0, self.max_calls - self.started_calls)
+
+    @property
+    def is_cancelled(self) -> bool:
+        if not callable(self._cancel_check):
+            return False
+        try:
+            return bool(self._cancel_check())
+        except Exception:
+            logger.debug("compression provider-work cancellation check failed", exc_info=True)
+            return False
+
+    def try_start_call(self) -> bool:
+        if self.is_cancelled or self.started_calls >= self.max_calls:
+            return False
+        self.started_calls += 1
+        return True
+
+    def finish_call(self) -> None:
+        self.completed_calls += 1
+
+
+_COMPRESSION_ATTEMPT_PROVIDER_BUDGET: contextvars.ContextVar[
+    Optional[_CompressionAttemptProviderBudget]
+] = contextvars.ContextVar("hermes_compression_attempt_provider_budget", default=None)
+
+
+@contextlib.contextmanager
+def _compression_attempt_provider_budget(
+    *,
+    cancel_check=None,
+    max_calls: int = _LEAN_ATTEMPT_MAX_PROVIDER_CALLS,
+):
+    """Bound one synchronous compressor attempt without cross-worker state."""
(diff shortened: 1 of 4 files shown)


Enough1122:

Reviewed `agent/context_compressor.py`, `agent/conversation_compression.py` + tests for PR #97671. The approach is solid: a context-local (ContextVar) provider-call budget scoped to one compression attempt, with `compress()` wrapping `_compress_impl()` so every path is bounded, and commit-fence cancellation propagated into in-flight auxiliary work. The `_cancel_check` forwarding guard in `_supported_compression_kwargs` (only for an explicitly declared `_cancel_check` param, never via `**kwargs`) is a nice touch that keeps plugin engines clean.

Non-blocking observations:

1. **Shared budget can starve the summary if digests ever run first.** `_build_chunk_digests` caps itself at `budget.remaining_calls` and will consume every remaining slot (up to 7 after the summary). `_generate_summary` then calls `_start_compression_provider_call()`, raising `_CompressionAttemptBudgetExhausted` and returning `None` (summary silently skipped) when no slot is left. The new test `test_remaining_attempt_budget_coarsens_all_digest_input` explicitly reserves the main-summary request before digests (`assert budget.try_start_call()  # reserve the main-summary request`), which signals the intended summary-first ordering — but nothing in the budget code enforces it. If any lean path dispatches digests before the summary reservation, the attempt loses its summary with only a warning log. Worth an explicit ordering guarantee (or reserving the summary slot at attempt start) rather than relying on call order.

2. **`_CompressionAttemptProviderBudget.completed_calls` is never read.** `finish_call()` increments it and `_finish_compression_provider_call()` calls it in `finally`, but no decision uses `completed_calls` — budget admission is entirely `started_calls`-based. Harmless, but it's dead state that suggests an intended-but-missing use (e.g. refunding slots for failed calls). Either drop it or wire it into `remaining_calls`.

3. **Coarser digests may exceed the per-chunk digest cap.** Capping at 8 provider calls (1 summary + 7 digests, vs. the previous up-to-28 chunk limit) means each segment now covers ~4× the text. The `_LEAN_DIGEST_MAX_TOKENS = 1_400` per-chunk cap was sized for ~13:1 on the default 72k-char chunk; with coarser segmentation the ratio to actual source grows and a digest could approach the model's output limit or lose detail. The trade-off is intentional per the PR body, but a guard on per-segment size (or accepting a coarser output marker like `[digest unavailable …]`) would make the behavior explicit.

4. **`_start_compression_provider_call` is silent when unbudgeted.** Outside a budget context it always returns `True`, which preserves old behavior for direct callers of `_generate_summary`/`_build_chunk_digests`, but it also means the "8-call budget" only holds when the caller goes through `compress()`. A brief docstring noting the unbounded fallback would help future callers.

The fence-cancel-into-in-flight test (`test_compression_fence_cancel_is_not_masked_by_unset_hard_cancel_event`) and the `_supported_compression_kwargs` opt-in test are good coverage. Nothing blocking.

fangliquanflq:
Thanks for the detailed review. I traced each observation against the full compression call graph:

1. No ordering change was needed. The only non-recursive production path calls `_generate_summary()` from `_compress_impl()` before `_augment_summary_lean()` can dispatch digests. The deterministic fallback can dispatch digests without a summary reservation, but it has no later summary call to starve.
2. Valid and addressed: removed `completed_calls`, `finish_call()`, and the corresponding completion bookkeeping. Admission remains intentionally based on started calls, so failed provider calls still consume the attempt budget.
3. This is the intentional bounded-cost trade-off. Coarser segments still have the existing 1,400-token output cap, so they cannot exceed that cap; exact identifiers and recovery remain covered by the deterministic anchor index, verbatim user section, recovery footer, and `session_search` pointers.
4. Valid and addressed: `_start_compression_provider_call()` now documents that direct callers outside `compress()` retain legacy unbounded behavior.

The follow-up review also found that cancellation between digest calls needed to remain distinct from ordinary budget exhaustion. Cancellation now propagates as `AuxiliaryExplicitCancellation` before compressor state publication, while true budget exhaustion still degrades normally. The regression test verifies cancellation after the first digest stops the attempt rather than returning a partial result.

Validation:
- `scripts/run_tests.sh tests/agent/test_compression_attempt_ownership.py tests/agent/test_auxiliary_explicit_cancellation.py -q` — 28 passed
- `scripts/run_tests.sh tests/agent/test_context_compressor.py tests/agent/test_compression_concurrent_fork.py tests/agent/test_compression_worker_isolation_76354.py tests/agent/test_compression_attempt_ownership.py -q` — 212 passed
- Merge-tree against current upstream `main` and `git diff --check` both passed

poisdahl:
The eight-call attempt budget is useful and the digest coarsening preserves source coverage, but the PR description currently overstates the in-flight cancellation guarantee.\n\nThe cancel_check is consumed by _run_protected_sync_provider_call. On cancellation that seam returns the compression owner promptly by raising AuxiliaryExplicitCancellation, while callback(kwargs) continues in a daemon thread until its provider timeout. I verified the lifetime on current main plus the related total-deadline composition: with a blocking callback and a 50 ms fence deadline, the owner returned with AuxiliaryExplicitCancellation while the provider callback remained alive and finished only after it was explicitly released.\n\nFor this PR, that means:\n- the old owner cannot dispatch another digest, which is fixed;\n- the attempt budget caps logical call_llm starts made by that owner, which is fixed;\n- the already-started provider request is not necessarily stopped;\n- repeated cancelled attempts can still accumulate one live provider callback apiece after their owners and ordinary compression leases unwind.\n\nThis is the same protected-provider seam discussed on #97064 and in my concrete #97512 review: https://github.com/NousResearch/hermes-agent/pull/97512#issuecomment-5468789776\n\nI would keep the budget change, but narrow the in-flight claim unless the implementation also closes or cooperatively cancels each transport stream. A regression should assert actual provider_done before retry admission, not only owner completion. If a transport cannot be interrupted, the detached callback needs a process-wide bound whose slot remains occupied until the callback really exits.\n\nReview setup: exact PR head composed cleanly with current main 66666f6e; CI is green. I reviewed the budget and cancellation paths end to end and ran the focused callback-lifetime probe, but did not run the full suite locally.

teknium1:
Note for rebase: PR #98628 (merged, 4f22543509) collapsed lean compaction to a single auxiliary request per attempt, which removes the multi-call digest pipeline this PR's scenarios multiply through. The attempt-lifecycle concern (ceiling/supersession across ATTEMPTS, cf. #97488) may still stand — please rebase and re-scope to what remains.

fangliquanflq:
Thanks — confirmed against current `main`. PR #98628 removes `_build_chunk_digests` and makes lean compression exactly one auxiliary call, so this PR's call-budget and digest-coarsening changes no longer apply.

I also rechecked the residual fence-cancellation change before attempting to preserve it. `_run_protected_sync_provider_call()` deliberately lets the owner unwind while the daemon provider callback continues, while `call_llm()` releases the per-task semaphore in the owner's `finally`. Keeping fence cancellation alone would therefore allow a later compression request to exceed `auxiliary.compression.max_concurrency` without stopping the in-flight wire request. The existing commit fence already prevents that late result from being published.

There is consequently no safe residual code patch from this branch to push after #98628. The remaining attempt-lifecycle work in #97488 needs provider/permit lifetime coupling (or cooperative transport closure), rather than this PR's former digest-attempt budget. No code was pushed; the attempted merge was aborted cleanly after confirming the conflict and superseded scope.

Issue #8512, opened 2026-04-12
[Bug]: BlueBubbles webhook fails on macOS due to IPv6 localhost resolution


BukeLy:
### Description

On macOS, Node.js (BlueBubbles server) resolves `localhost` to IPv6 `::1` by default. The Hermes gateway (aiohttp) binds to IPv4 `127.0.0.1` only. When the webhook URL uses `localhost`, all webhook deliveries fail silently with `ECONNREFUSED ::1:<port>`.

### Steps to Reproduce

1. Start Hermes gateway with BlueBubbles platform on macOS
2. Gateway registers webhook URL as `http://localhost:8645/bluebubbles-webhook`
3. BlueBubbles resolves `localhost` to `::1` (IPv6)
4. Webhook delivery fails: `Error: connect ECONNREFUSED ::1:8645`

### Expected Behavior

Webhook URL should use `127.0.0.1` (explicit IPv4) so BlueBubbles connects to the correct address.

### Actual Behavior

All inbound iMessages are silently dropped. No error in Hermes logs (the webhook never reaches Hermes).

BlueBubbles server log:
```
[WebhookService] Failed to dispatch "new-message" event to webhook: http://localhost:8645/...
  -> Error: connect ECONNREFUSED ::1:8645
```

### Affected Component

Gateway - BlueBubbles platform adapter

### Platform

macOS 15 (Sequoia), BlueBubbles 1.9.9

### Hermes Version

v0.8.0

kotafarms:
Confirming repro on macOS 26.3 (Apple Silicon) with BlueBubbles Server v1.9.9.

Webhook listener binds to `127.0.0.1:8645` but Hermes registers the URL as `http://localhost:8645/...`. macOS resolves `localhost` → `::1` (IPv6) first, and BlueBubbles' Node HTTP client doesn't fall back to IPv4 → `ECONNREFUSED ::1:8645` on every `new-message` event.

Worked around locally by changing the loopback rewrite in `gateway/platforms/bluebubbles.py::_webhook_url` from `"localhost"` → `"127.0.0.1"` (the same one-liner already proposed in PRs #21560 and #23998). Confirmed end-to-end iMessage round-trip works after the patch.

Would be great to get one of those PRs merged — happy to retest against whichever lands.

tmchow:
I just hit this as well. came to report it and found this issue.

laurinaitis:
Two things that apply to every fix open on this, gathered while carrying one of them locally.

[...]

PR #8263, opened 2026-04-12
fix(bluebubbles): use 127.0.0.1 instead of localhost for webhook URL

BukeLy:
## What does this PR do?

On macOS, Node.js (BlueBubbles server) resolves `localhost` to IPv6 `::1` by default. The Hermes gateway (aiohttp) binds to IPv4 `127.0.0.1` only. When the webhook URL uses `localhost`, all webhook deliveries fail silently with `ECONNREFUSED ::1:<port>`, so no inbound iMessages are ever received.

This PR normalizes loopback webhook hosts to `127.0.0.1` (explicit IPv4) instead of `localhost`, ensuring BlueBubbles connects to the correct address.

## Related Issue

Fixes #8512

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `gateway/platforms/bluebubbles.py`: Changed `_webhook_url` property to normalize local hosts to `127.0.0.1` instead of `localhost`
- `tests/gateway/test_bluebubbles.py`: Updated `TestBlueBubblesWebhookUrl` tests to verify IPv4 normalization

## How to Test

1. Start Hermes gateway with BlueBubbles platform on macOS
2. Verify webhook registers with `http://127.0.0.1:...` instead of `http://localhost:...`
3. Send an iMessage → confirm webhook delivery succeeds and gateway processes the message

**BlueBubbles server log (before fix):**
```
[WebhookService] Failed to dispatch "new-message" event to webhook: http://localhost:8645/...
  -> Error: connect ECONNREFUSED ::1:8645
```

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
- [x] My PR contains **only** changes related to this fix/feature (no unrelated commits)
- [x] I've run `pytest tests/ -q` and all tests pass
[...]

PR #97672, opened 2026-08-29
fix(bluebubbles): preserve IPv4 webhook address

Burry:
## Summary
- preserve an explicitly configured IPv4 loopback callback address
- normalize only wildcard bind addresses for BlueBubbles webhook registration
- add regression coverage for default and wildcard hosts

## Validation
- focused runtime assertions for webhook URL and Clu mention matching
- python -m py_compile gateway/platforms/bluebubbles.py
- git diff --check
- live BlueBubbles webhook re-registered at 127.0.0.1; missed mention replayed and response delivered

The local environment does not currently include pytest, so the focused pytest selection could not be executed.

### DIFF
--- a/gateway/platforms/bluebubbles.py
+++ b/gateway/platforms/bluebubbles.py
@@ -340,7 +340,11 @@ class BlueBubblesAdapter(BasePlatformAdapter):
     def _webhook_url(self) -> str:
         """Compute the external webhook URL for BlueBubbles registration."""
         host = self.webhook_host
-        if host in {"0.0.0.0", "127.0.0.1", "localhost", "::"}:
+        # Wildcard bind addresses are not valid callback destinations. Preserve
+        # an explicit IPv4 loopback address, though: rewriting 127.0.0.1 to
+        # ``localhost`` lets some Node runtimes prefer ::1 while aiohttp is
+        # listening only on IPv4, causing silent webhook delivery failures.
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: #8263 and #8512. This PR preserves an explicitly configured IPv4 callback, while #8263 normalizes every local callback form to IPv4. They overlap but use different normalization policies.

krullma:
### E2E evidence from a real deployment (macOS, gateway + BlueBubbles Server on the same host)

Current verified state:
- BlueBubbles webhook registry contains exactly one entry: `http://127.0.0.1:8645/bluebubbles-webhook` (literal IPv4, confirmed via `GET /api/v1/webhook`).
- Gateway listener: aiohttp bound to IPv4 `127.0.0.1:8645` only (`lsof` confirms; `curl http://[::1]:8645/health` fails while `http://127.0.0.1:8645/health` succeeds).

Failure mode previously hit on this same deployment: when the registered URL was `http://localhost:8645/...`, BlueBubbles' dispatcher POSTed events to `::1:8645` (macOS resolves localhost → ::1) and got `ECONNREFUSED` (visible in bluebubbles-server main.log), so inbound iMessage events silently vanished while outbound REST kept working. Re-registering the literal `127.0.0.1` URL restored inbound delivery.

This PR's behavior (preserve the IPv4 address instead of rewriting to `localhost`) matches what the deployment needs.

PR #97675, opened 2026-08-29
fix(mcp): make reconnect retries safe for write tools

jaxleezhang:
## Summary
- Read paths (`list_resources`, `read_resource`, `list_prompts`, `get_prompt`) now do a bounded wait for a fresh session on `session expired`/`session not found` instead of failing immediately, then retry once.
- Only tool calls that are known reads (discovery-time `readOnlyHint is True`) are transparently replayed after a reconnect.
- Write or unknown-hint tool calls do not get silently replayed after a session-expiry reconnect — they return an `uncertain_outcome` result instead, so the caller can decide how to proceed rather than risking a duplicate side effect.
- This does not claim exactly-once semantics for the underlying third-party MCP server; it only avoids the agent itself blindly replaying a call whose original outcome is unknown.

## Test plan
- [x] `tests/tools/test_mcp_tool_session_expired.py`
- [x] `tests/tools/test_mcp_reconnect_bounded_wait.py`
- [x] `tests/tools/test_mcp_session_expired_write_no_replay.py`
- 27 passed, 0 failed (run via `uv run python -m pytest` on the three files above)

Originally surfaced against a Google Calendar MCP server (session expiry mid-tool-call), but the fix is a general MCP reconnect/replay-safety change, not calendar-specific.

### DIFF
--- a/tests/tools/test_mcp_reconnect_bounded_wait.py
+++ b/tests/tools/test_mcp_reconnect_bounded_wait.py
@@ -1,0 +1,132 @@
+"""Tests for bounded readiness waits on MCP read-path handlers.
+
+When a call lands while an MCP server is still completing its initial
+connection or a reconnect, ``server.session`` is briefly ``None``. The
+``tools/call`` handler (``_make_tool_handler``) already waits briefly for a
+fresh session instead of failing immediately (see #26892). The four
+protocol-level read handlers -- ``list_resources``, ``read_resource``,
+``list_prompts``, ``get_prompt`` -- did not share that wait: they returned
+"not connected" the instant ``session`` was ``None``, even though a
+reconnect might complete a fraction of a second later. This file locks in
+the fix: all read-path handlers wait up to a bounded timeout for a live
+session before giving up, and still fail (rather than hang) when the
+session never comes back.
+"""
+import json
+import threading
+import time
+from unittest.mock import MagicMock
+
+import pytest
+
+
+def _install_connecting_server(name: str):
+    """Register a real MCPServerTask with no session yet (mid-connect)."""
+    from tools import mcp_tool
+    from tools.mcp_tool import MCPServerTask
+
+    mcp_tool._ensure_mcp_loop()
+    server = MCPServerTask(name)
+    server.session = None
+    mcp_tool._servers[name] = server
+    return server
+
+
+def _make_ready_result(session_method: str):
+    """Return a MagicMock session whose *session_method* succeeds cleanly."""
+    session = MagicMock()
+
+    if session_method == "list_resources":
+        result = MagicMock(resources=[], nextCursor=None)
+    elif session_method == "read_resource":
+        result = MagicMock(contents=[])
+    elif session_method == "list_prompts":
+        result = MagicMock(prompts=[], nextCursor=None)
+    else:  # get_prompt
+        result = MagicMock(messages=[], description=None)
+
+    async def _call(*a, **kw):
+        return result
+
+    setattr(session, session_method, _call)
+    return session
+
+
+@pytest.mark.parametrize(
+    "handler_factory, handler_kwargs, session_method, call_args",
+    [
+        ("_make_list_resources_handler", {"tool_timeout": 2.0}, "list_resources", {}),
+        ("_make_read_resource_handler", {"tool_timeout": 2.0}, "read_resource", {"uri": "file://x"}),
+        ("_make_list_prompts_handler", {"tool_timeout": 2.0}, "list_prompts", {}),
+        ("_make_get_prompt_handler", {"tool_timeout": 2.0}, "get_prompt", {"name": "p1"}),
+    ],
+)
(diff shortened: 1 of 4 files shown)


Enough1122:

Reviewed `tools/mcp_tool.py` + tests for PR #97675. The core design is correct and important: a session-expired error can fire *after* a write reached the remote server, so reconnecting-but-not-replaying for non-read-only tools (with an `uncertain_outcome` error + verify-state guidance) is the right call. Keeping the transport heal unconditional (reconnect still happens even when replay is skipped) is a nice touch, and the bounded-wait for the read-path handlers closes a real gap (#26892 parity). Test coverage is thorough (readOnlyHint gating parametrized over True/False/None, non-replay, bounded timeout).

Non-blocking observations:

1. **Fail-closed hint may silently disable auto-replay for genuinely read-only tools.** Replay is now gated on `_tool_read_only_hints[server][tool] is True`, and `_annotation_read_only_hint` returns True only when the *server* self-declares `annotations.readOnlyHint: true` at discovery. Many MCP servers don't emit that annotation, so their read tools (e.g. `list_resources`, `read_resource` on a server with no readOnlyHint) now return `uncertain_outcome` on a transient session-expired instead of auto-replaying and succeeding — a behavior change vs. before, where every session-expired call was replayed once. Evidence this isn't hypothetical: the existing `get_bank`/`health` read tests needed explicit manual hint markers (`_tool_read_only_hints[...] = {"get_bank": True}`) to preserve the old replay behavior. Worth confirming the discovery path annotates reads reliably in practice, or at least monitoring for an uptick in `uncertain_outcome` on read calls after this ships. (The fail-closed default is the correct security posture for the *write* case — just flagging the read-path blast radius.)

2. **`_bump_server_error` on the non-retryable path counts a healed transport toward the circuit breaker.** In the `retryable=False` branch, the transport was successfully reconnected but `_bump_server_error(server_name)` still increments the consecutive-failure count — three such session-expired write events in a row opens the breaker (`_CIRCUIT_BREAKER_THRESHOLD = 3`) for the 60s cooldown even though the server itself is healthy. Since the reconnect succeeded, consider resetting (or not bumping) the error count on that path, so write-outcome-uncertainty doesn't take down subsequent reads.

3. **Read-handler wait cap differs from the tool-handler wait.** `_get_ready_server_for_read` uses `min(5.0, tool_timeout or 5.0)`, while `_wait_for_server_session_ready` defaults to 15s and `_make_tool_handler`'s session wait uses the tool timeout. A server that takes >5s to reconnect will therefore report "not connected" to read calls even when a larger `tool_timeout` is configured and the tool path would wait longer. Likely fine (5s is a reasonable bounded wait), just an inconsistency worth a comment or a shared helper so the two paths don't drift.

The uncertainty surfaced to the agent (reconnected, outcome unknown, verify state) is good UX for the write case; nothing blocking.

jaxleezhang:
Thanks — this is a careful read. Taking the three in order.

**2 is a real bug and I will fix it.** The non-retryable path calls `_bump_server_error()` after the reconnect *succeeded*, which contradicts the contract `_reset_server_error` documents in its own docstring: "Call this on any unambiguous success signal (successful tool call, successful reconnect, manual /mcp refresh)." A successful reconnect is exactly that signal. As written, three session-expired writes in a row open the breaker for 60s on a server that is demonstrably healthy — and the breaker is per-server, so it then blocks reads that would have worked. Switching that path to `_reset_server_error(server_name)` is the right call: the write outcome is uncertain, but the *transport* is not, and the breaker tracks transport health. Will push that with a test asserting the breaker stays closed across three consecutive non-replayed writes.

**1 — agreed on the substance, with one correction to the example.** `list_resources`/`read_resource` are the protocol-level `resources/*` handlers; they call `_handle_session_expired_and_retry` with the default `retryable=True` and are not gated by the hint, so those specific calls keep replaying as before. The gate only applies to `tools/call` in `_make_tool_handler`.

The concern still stands where it matters, though: a genuinely read-only *tool* on a server that never emits `annotations.readOnlyHint` now returns `uncertain_outcome` instead of transparently replaying. That is a deliberate trade — `_annotation_read_only_hint` is fail-closed by design, and the failure mode it prevents (a duplicated calendar event, a double-charged write) is much worse than the one it introduces (an extra verify round-trip on a read). The test markers you spotted are the honest evidence of the blast radius, not a workaround. I would rather not widen the gate with heuristics like name prefixes — that reintroduces exactly the guessing this PR removes. Monitoring `uncertain_outcome` on the read path after ship sounds right; if it is noisy, the fix belongs upstream in discovery (encouraging servers to annotate) rather than in the retry logic.

**3 — fair, will align.** The `min(5.0, tool_timeout or 5.0)` cap versus `_wait_for_server_session_ready`'s 15s default is an inconsistency I introduced without meaning to. I will pull the bound into a named constant next to the other reconnect timings so the two paths cannot drift silently, and keep the read cap conservative.

Separately: the branch has gone to conflict against main since I opened this. I will rebase along with the above.

PR #96851, opened 2026-08-28
fix(gateway): stop blocking the event loop — off-loop hot sites + ASYNC lint ratchet

kshitijk4poor:
## Summary

Architectural fix for the **event-loop blocking** bug class (Pattern A): blocking calls inside `async def` freeze the whole gateway/uvicorn event loop — every adapter, timer, health check, and concurrent chat stops until the call returns.

This class keeps producing individual incident PRs — a 17-minute `getaddrinfo` freeze took the backend down (#91912), `list_active_loops()` froze the loop (#92413), media cache writes (#93967), `atomic_json_write` on adapter hot paths (#84168), credential resolution (#85017, #84891). Each fix lands and the next blocking call re-enters unnoticed. This PR fixes the remaining unguarded core sites **and adds the CI gate that stops the class from re-entering.**

## Real-world impact

**WHO:** anyone running the gateway (Telegram/Discord/Slack/webhook users) or the Desktop backend.
**WHEN:** a webhook fires a `gh` network call, someone runs `/save` on a long session, voice TTS reads a large audio file, or a gateway `--replace` restart waits for the old PID.
**WHY it matters:** during those windows the event loop is frozen — messages sit undelivered on every platform, health checks time out (systemd/launchd can restart-loop the service), and streaming stalls.

## Fixes (4 sites)

| Site | Before | After |
|---|---|---|
| `gateway/platforms/webhook.py` `_deliver_github_comment` | `subprocess.run(gh …, timeout=30)` inline — froze the loop up to 30s per delivery | `await asyncio.to_thread(subprocess.run, …)` |
| `gateway/run.py` `start_gateway --replace` | `time.sleep(0.5)`×20 + `time.sleep(0.25)`×20 — up to 15s frozen during restart | `await asyncio.sleep(…)` (re-lands #36163 at current line numbers — credit @AhmetArif0, Co-authored-by preserved) |
| `gateway/slash_commands.py` `/save` | session render + file write inline (scales with transcript size, multi-MB) | one `asyncio.to_thread` hop for render+write |
| `hermes_cli/web_server.py` voice TTS | multi-MB audio read + unlink inline on uvicorn loop | `asyncio.to_thread`, unlink rides the same hop |

## Prevention gate (the architectural part)

[...]

PR #95380, opened 2026-08-26
[Perf]: list_sessions_rich re-walks compression chains N+1 per row under the global write lock — redundant with its own recursive CTE

Finn763:
Closes #95316

## Summary

`SessionDB.list_sessions_rich()` resolved each compression root's tip by re-walking its chain hop-by-hop through `get_compression_tip()` — **one query per hop per row**, taken under the store's global lock whenever no pooled WAL reader is available — redundant with the recursive `chain`/`chain_max` CTE the same method already runs on the `order_by_last_active=True` path. This replaces that N+1 walk with a single batched statement + in-memory traversal:

- New `_compression_child_edges()`: lifts `get_compression_tip()`'s per-hop child query over all compression-ended parents in ONE statement — identical child filters (`_branched_from` / `_delegate_from` / tool children excluded) and identical preference ORDER BY (continuing chain → still-live → closed; freshest first; higher id breaks ties). Read through `_read_ctx()` so it borrows a pooled WAL reader instead of checking out one reader per hop.
- New `_resolve_compression_tips(root_ids)`: builds the edge index once, then walks every root's chain entirely in memory with `get_compression_tip()`'s exact hop semantics (best-first child, cycle guard, 100-hop bound). A drop-in for the per-hop walker.
- The stale in-code comment ("per-session graph walk, not batchable in one query") is gone along with the loop it justified.

Sidebar/session-picker reads stop paying R×D read-pool checkouts per listing; in the degraded path (no pooled reader — non-WAL / open failure / pool ceiling, where `_read_ctx()` falls back to the writer under `self._lock`) they stop paying R×D lock acquisitions too.

## Benchmark

`scripts/benchmark_list_sessions_rich.py [roots] [depth]` (synthetic store: R compression roots × D-deep chains + R live fillers, whole store on one page, best of 9 reps):

**120 roots × depth 8 (+120 fillers):**

| metric | before | after |
|---|---|---|
| `order_by_last_active=True` | 19.66 ms | 14.13 ms |
| `order_by_last_active=False` | 16.40 ms | 10.83 ms |
| `get_compression_tip()` calls | 120 | **0** |
| SQL statements per listing | 1082 | **3** |

**40 roots × depth 5 (+40 fillers):**

| metric | before | after |
|---|---|---|
| `order_by_last_active=True` | 5.79 ms | 4.39 ms |
| `order_by_last_active=False` | 4.85 ms | 3.55 ms |
| `get_compression_tip()` calls | 40 | **0** |
[...]

PR #90734, opened 2026-08-20
fix(state): unlocked reads on the shared SessionDB writer connection destroy turns as session_persistence_failed

Kyzcreig:
## Symptom
A live gateway turn (17 API calls of real work) died at persist time with:
```
WARNING run_agent: Session DB append_message failed: <TrackedConnection ...> returned NULL without setting an exception
WARNING agent.conversation_loop: Turn ended: reason=session_persistence_failed
```
The user saw *"No reply: session storage could not be written"* — while disk, permissions, and the database were all healthy.

## Root cause (reproduces on current main)
`SessionDB` shares ONE writer connection (`check_same_thread=False`) guarded by `self._lock`. Four read-only methods execute statements on that same connection object **without taking the lock**:

| method | live caller |
|---|---|
| `get_compression_lock_holder` | gateway `_session_has_compression_in_flight` via `asyncio.to_thread` (interrupt/busy-input hot path) |
| `list_pending_handoffs` | gateway handoff watcher poll |
| `get_handoff_state` | handoff flow |
| `clear_session_activity_labels` no-op fast path | every turn's `finally` |

When one of those SELECTs races a turn-boundary `append_messages_batch` on the same connection, CPython's sqlite3 layer raises a bare `SystemError` ("returned NULL without setting an exception"). `SystemError` is **not** a `sqlite3.Error`, so it escapes every branch of `_execute_write`'s retry net (locked/busy, no-more-rows, FTS rebuild) and the turn is destroyed.

**Deterministic reproduction:** 1 writer thread doing turn-shaped batch appends + 3 reader threads on any one of the four methods → the exact error within seconds; writer alone is clean. The repro runs are the regression tests in this PR, and all of them fail on unmodified `main` (5 of 7 RED pre-fix).

## Fix — two independent layers
1. **Route the four reads through `_read_ctx()`** (which exists for exactly this): under WAL that's a per-thread read-only connection — no lock needed, structurally cannot race the writer; on the non-WAL fallback it's the shared connection under `self._lock`, byte-for-byte the legacy behavior.
2. **Defense in depth:** `_execute_write` retries the message-scoped `SystemError` like locked/busy (same patience deadline, warning logged), so any future reintroduction of the race degrades to a retried write instead of a destroyed turn. Any other `SystemError` propagates untouched on attempt 0.

## Tests
`tests/state/test_writer_conn_thread_safety.py` — 7 tests, RED-first proven on clean main (5 fail pre-fix / 7 pass post-fix):
[...]

PR #82428, opened 2026-08-09
fix(state): isolate background reads from writer connection

milnerrad:
## Summary

- Route four read-only `SessionDB` paths through `_read_ctx()` instead of directly using the shared writer connection.
- Preserve per-thread WAL readers and the existing locked fallback for non-WAL/read-connection failures.
- Add a regression that poisons `_conn` and proves background state reads never touch it.

## Problem

`SessionDB._conn` is opened with `check_same_thread=False` and is shared by gateway worker threads. Most access is serialized by `self._lock`, but four production reads bypassed both `_read_ctx()` and the lock:

- `get_compression_lock_holder()`
- `clear_session_activity_labels()` fast path
- `get_handoff_state()`
- `list_pending_handoffs()`

The always-running handoff watcher calls `list_pending_handoffs()` through `AsyncSessionDB` every two seconds, while transcript appends prepare statements on the same writer connection.

CPython 3.11 releases the GIL around `sqlite3_prepare_v2()`. If one prepare fails while another thread successfully uses the same connection, the successful call can replace SQLite's connection-global error state before `_pysqlite_seterror()` reads it. CPython then returns NULL without a Python exception and surfaces:

```text
<hermes_cli.sqlite_safe_read.TrackedConnection object ...>
returned NULL without setting an exception
```

Hermes treats that as a session persistence failure and rolls back the turn. This was reproduced on Python 3.11.15 / SQLite 3.53.1 by concurrently preparing successful and failing statements on one `check_same_thread=False` connection.

This addresses the underlying shared-connection race rather than broadly retrying every `SystemError` as proposed in #77045.

## Fix

All four reads now use `_read_ctx()`:

- WAL: per-thread read-only connection, no writer contention.
- DELETE mode or read-connection failure: shared writer under `self._lock`.

An AST audit found no remaining production `self._conn` access outside initialization or methods holding `self._lock`; the private FTS probe is called only under its public maintenance methods' lock.

## Verification

- RED: new regression failed on current `main` with `AssertionError: read touched shared writer connection`.
- GREEN: regression passes after the fix.
- `295 passed` across state, FTS, write-lock, compression-lock, handoff, activity, and read-path suites.
- Stress check: 1,000 transcript writes with four concurrent background-reader threads; `0` errors.
- Ruff passed.
- Python compilation passed.
- `git diff --check` passed.
- Independent review found no security concerns or logic errors.

## Risk

[...]

PR #97676, opened 2026-08-29
perf(state): route 39 pure-read SessionDB methods off the writer lock + AST gate

kshitijk4poor:
## Summary

Pattern-C architectural fix (SQLite write-lock contention) — the third gate in the series with #96851 (Pattern A: event-loop blocking → ruff ASYNC ratchet) and #96922 (Pattern B: O(N²) hot paths → scaling guards).

Completes what #90734 started. That PR fixed the four **unlocked** readers racing the shared writer connection (a crash class). This PR fixes the complementary half: **39 pure-read methods that correctly took `self._lock` — and therefore convoyed every concurrent turn's persistence behind each read** (a contention class), plus the CI gate that stops the pattern from re-entering.

## Real-world impact

**WHO/WHEN:** every gateway install — the gateway shares ONE `SessionDB` across all agents. Any of these reads (session counts for the sidebar, titles, telegram topic bindings, compression tips, meta reads, prune scans) blocks ALL turn persistence for its duration, and vice versa: readers stall behind multi-MB `append_messages_batch` flushes.
**WHY `_read_ctx()`:** under WAL it serves reads from a bounded pool of read-only connections with **no lock**; on non-WAL installs it falls back to the locked writer **byte-for-byte** — so DELETE-journal installs are exactly as before.

## Measured (2 writer + 3 reader threads, 3s, WAL)

| | reads served | p99 | max |
|---|---|---|---|
| before (locked readers) | 78.5k | 2.49ms | 160.7ms |
| after (`_read_ctx`) | **200.4k (2.6x)** | **0.53ms (4.7x)** | **25.5ms (6.3x)** |

Writes unchanged (1.5k vs 1.4k — noise).

**Honest caveat:** the currently-bundled SQLite (3.46) trips the WAL-reset-vulnerability gate (#70055) → new DBs run DELETE journal, where `_read_ctx()` degrades to the identical locked path and this change is **behavior-neutral by construction**. The win applies to WAL installs today (legacy WAL DBs, `database.journal_mode` operators, fixed runtimes) and to everyone once the bundled runtime upgrades. The conversion is correct in both worlds; it simply waits for WAL where WAL is refused.

## The conversion (mechanical + audited)

`with self._lock:` → `with self._read_ctx() as conn:` and `self._conn.` → `conn.` in 39 SELECT-only method bodies. Audited three ways:
1. Per-method write-verb scan (INSERT/UPDATE/DELETE/REPLACE/executemany/_execute_write) — read-modify-write methods (`set_session_archived` loops, cooldown restore, FTS fail-open, checkpoint PRAGMAs) **stay on the writer lock**.
2. AST placement check — every `conn` use in all 39 methods is inside its with-block (no escaped cursor).
3. The gate itself (below) — zero violations on this branch, 38 on pre-conversion main.

## The gate (the architectural part)

`tests/state/test_no_locked_readers_gate.py` — an AST scanner over `SessionDB` that fails CI when any method holds `with self._lock:` around a **pure-read** body (SELECT/read-PRAGMA only, no write calls). Write-PRAGMAs (`wal_checkpoint`, `optimize`, …) and any `self.<write-ish>()` call under the lock legitimately pass. Sabotage self-check included: a synthetic locked reader must be flagged, a locked writer must not.

Same ratchet philosophy as #96851's ASYNC gate: the class of regression is greppable/AST-detectable, so it gets a structural gate, not another one-off fix.

## Verification

- `tests/test_hermes_state.py`: 243 passed
- Combined state sweep: **540 passed** — the only 3 reds are the pre-existing `test_fts_runtime_rebuild` failures, verified failing on clean `origin/main` before this change
- Gate: 2/2 (zero violations + sabotage self-check)
- ruff green; rebased on fresh main; post-rebase 245 green

## Provenance

- Builds on the `_read_ctx()` infrastructure (#82428/#73803 lineage) rather than inventing a new mechanism.
- Complements (does not overlap) #90734 — different method sets; that PR's four unlocked readers are untouched here.
- #95380's batched compression-edge query composes cleanly (its `_read_ctx` reads are the same pattern).

### DIFF
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -6490,8 +6490,8 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
         if active_only:
             query += " AND ended_at IS NULL"
         query += " ORDER BY last_active DESC"
-        with self._lock:
-            rows = self._conn.execute(query, params).fetchall()
+        with self._read_ctx() as conn:
+            rows = conn.execute(query, params).fetchall()
         return [self._session_row_dict(r) for r in rows]
 
     def find_session_by_origin(
@@ -6524,8 +6524,8 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
             query += " AND COALESCE(thread_id, '') = ?"
             params.append(str(thread_id))
         query += " ORDER BY started_at DESC"
-        with self._lock:
-            rows = [dict(r) for r in self._conn.execute(query, params).fetchall()]
+        with self._read_ctx() as conn:
+            rows = [dict(r) for r in conn.execute(query, params).fetchall()]
         if not rows:
             return None
         if user_id:
@@ -7599,8 +7599,7 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
         """Return the persisted deterministic-fallback streak."""
         if not session_id:
             return 0
-        with self._lock:
-            conn = self._conn
+        with self._read_ctx() as conn:
             if conn is None:
                 return 0
             row = conn.execute(
@@ -7680,8 +7679,7 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
         """
         if not session_id:
             return 0
-        with self._lock:
-            conn = self._conn
+        with self._read_ctx() as conn:
             if conn is None:
                 return 0
             row = conn.execute(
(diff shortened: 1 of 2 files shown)


kshitijk4poor:
**Formal gate results (folded as `112baae` — the follow-up commit):**

The /simplify-code pass (2 reviewers) found the v1 scanner had blind spots, and the gate got stronger for it:

1. **Five more locked readers were hiding** from the v1 scanner: `get_compression_fallback_streak` / `get_compression_ineffective_count` (behind `conn = self._conn` aliasing) and `list_gateway_sessions` / `find_session_by_origin` / `search_sessions` (behind SQL held in variables — the v1 scanner required provable-SELECT to flag). All five now converted. The scanner tracks aliases and flags no-proven-write lock blocks instead of silently skipping them; the sabotage self-check pins all three detection classes plus a mixed variable-SQL writer that must NOT fire. **Total: 44 methods converted; strengthened scanner reports 43 violations on pre-conversion main, zero here.**

2. **`get_meta` reverted to the writer lock** — its inline comment on main documents a real read-your-writes dependency (`fts_rebuild_step` reads rebuild progress a pooled WAL reader can't see mid-transaction). The blanket conversion had overridden a documented design decision; it's now the single justified allowlist entry, replacing the dead `_enter_fts_fail_open` entry.

3. **Deadlock audit (both reviewers + lead, independently):** zero converted methods are reachable while `self._lock` is held (12 remaining lock blocks, 83 `_execute_write` callbacks checked) — the non-WAL fallback's lock re-acquisition can never nest.

Mutation check on the final stack: prod file reverted to main → gate fails with 43 violations; restored → 245 green. Rebased, pushed.

PR #94547, opened 2026-08-25
fix(teams): authenticate Bot Framework connector attachment downloads (pasted images 401)

Sibbern:
## What does this PR do?

Fixes silently dropped inline/pasted images in the Microsoft Teams platform adapter.

Teams delivers two different kinds of inbound attachments: file uploads (paperclip) carry **pre-authenticated SharePoint downloadUrls**, while pasted/inline images carry a contentUrl on **`smba.trafficmanager.net`** (Bot Framework connector), which **requires the bot's own bearer token**. The adapter fetches all attachments unauthenticated, so every pasted image fails with `401 Unauthorized` and is dropped — with only a log warning the user never sees:

```
WARNING hermes_plugins.teams_platform.adapter: [teams] Failed to cache image attachment:
Client error '401 Unauthorized' for url 'https://smba.trafficmanager.net/emea/<tenant>/v3/attachments/<id>/views/original'
```

The fix acquires a Bot Framework token via the client-credentials flow (using the adapter's existing `self._client_id/_client_secret/_tenant_id`) and attaches it **only** for Bot Framework hosts (`*.trafficmanager.net` / `*.botframework.com`). SharePoint and all other URLs behave exactly as before; SSRF and redirect guards are unchanged. Token-acquisition failure degrades gracefully (warning + unauthenticated fetch — same behavior as today, no new failure mode).

## Related Issue

No existing issue; found and verified on a live deployment.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `plugins/platforms/teams/adapter.py`:
  - Added `TeamsAdapter._get_botframework_token()` — client-credentials token for scope `https://api.botframework.com/.default`, cached until ~5 min before expiry
  - `_fetch_attachment_bytes()` — sends the bearer token when the attachment host ends with `trafficmanager.net` / `botframework.com`
  - Inbound `image/*` attachments with Bot Framework contentUrls now route through the authenticated `_fetch_attachment_bytes()` + `cache_media_bytes()` instead of the unauthenticated `cache_image_from_url()`

## How to Test

1. Run a Teams bot (personal scope) with `supportsFiles: true` in the app manifest
2. **Paste** an image into the bot chat (Ctrl+V into the compose box — not the paperclip upload) and send
3. Before this fix: gateway log shows the 401 warning above and the agent never receives the image. After: the image caches and is delivered to the agent (verified end-to-end on a live EMEA-tenant deployment, including the token-cache path on subsequent images)
[...]

PR #73685, opened 2026-07-28
fix(teams): authenticate inline image attachment downloads

zenplace-system:
## Problem

Images pasted or dragged into a Teams message do not arrive as SharePoint links. Their `contentUrl` points at the Bot Connector attachment service:

```
https://smba.trafficmanager.net/<region>/<tenant>/v3/attachments/<id>/views/original
```

That endpoint requires a Bot Framework bearer token. Both the image path (`cache_image_from_url`) and `_fetch_attachment_bytes` send plain unauthenticated GETs, so every pasted image comes back `401 Unauthorized` and is dropped with only a WARNING in the log:

```
WARNING hermes_plugins.teams_platform.adapter: [teams] Failed to cache image attachment:
  Client error '401 Unauthorized' for url
  'https://smba.trafficmanager.net/apac/<tenant>/v3/attachments/<id>/views/original'
```

Because the drop is silent from the agent's point of view, it goes on to answer about a picture it never received. In the case that surfaced this, a user pasted a screenshot of a broken diagram and asked what was wrong with it; the agent never saw the image and guessed.

`_fetch_attachment_bytes`' docstring assumes Teams attachments always carry pre-authenticated SharePoint URLs. That holds for the file-upload path, not for inline images.

## Fix

- Mint the same client-credentials token the outbound activity path already uses (`scope=https://api.botframework.com/.default`), cached until shortly before expiry so a burst of attachments on one message does not hit the STS once per file.
- Attach it **only** for hosts on `_ALLOWED_TEAMS_SERVICE_HOSTS`, so a tampered `contentUrl` cannot exfiltrate it. httpx drops `Authorization` on cross-origin redirects, and the existing SSRF guard plus `_ssrf_redirect_guard` still apply.
- Report dropped attachments in the message text instead of letting them vanish, so the agent can say it could not read the file rather than guessing. This covers the document and direct-URL branches too, which had the same silent-drop behaviour.

## Tests

7 new tests in `tests/gateway/test_teams.py`:

- connector-host detection, including a look-alike host (`smba.trafficmanager.net.evil.example`) and a plaintext-scheme URL
- bearer header present for connector URLs, absent for SharePoint URLs (and no token minted at all in that case)
- an inline image reaching the agent as a media URL
- a failed download surfacing in the message text, with the user's own words preserved alongside the notice

[...]

PR #72977, opened 2026-07-27
fix(teams): authenticate pasted image downloads

acepgh:
## Summary

- download protected Teams/Bot Framework image attachments through the adapter's authenticated attachment path
- acquire the bearer token from the initialized Teams SDK app and attach it only for allowlisted Bot Framework hosts
- cache downloaded image bytes with the correct image extension while preserving SSRF and token-exfiltration protections
- add regression coverage for authenticated Bot Framework downloads, non-Bot-Framework token isolation, and pasted-image event routing

## Root cause

Teams pasted images arrive as protected `smba.trafficmanager.net/.../v3/attachments/.../views/original` URLs. The image path used the generic unauthenticated `cache_image_from_url`, which returned HTTP 401 and left the agent with text only.

## Validation

- `uv run --extra dev python -m pytest tests/gateway/test_teams.py -o 'addopts=' -q` — 60 passed
- `uv run --extra dev --extra teams python -m pytest tests/gateway/test_teams.py tests/gateway/test_teams_pipeline_runtime_wiring.py tests/plugins/test_teams_pipeline_plugin.py tests/hermes_cli/test_teams_pipeline_plugin_cli.py -o 'addopts=' -q` — 90 passed
- `uv run --extra dev ruff check plugins/platforms/teams/adapter.py tests/gateway/test_teams.py` — passed
- live credentialed download of the previously failing Teams attachment — 215,594-byte PNG returned with a valid PNG signature
- restarted the Gilbert profile gateway; local `/health` returns `ok` and port 3985 is listening

## Security

The Bot Framework bearer token is only attached when the parsed hostname exactly matches the existing Bot Framework service-host allowlist. SharePoint and other pre-authenticated attachment URLs never receive the bot token. Existing SSRF URL and redirect guards remain in place.

PR #58476, opened 2026-07-04
fix(teams): authenticate pasted-image fetches, and restore emoji as Unicode

slvnlrt:
> **Updated 2026-08-19** — rebased onto current `main` (the previous revision was ~7000 commits behind) and **extended with a second, causally linked fix**: see *Emoji sprites* below. The authentication bug is still reproducible against `main` as of this rebase — `_fetch_attachment_bytes` sends no `Authorization` header and the image path still calls `cache_image_from_url(content_url)` bare.

## Relation to existing PRs

This bug overlaps with prior work — flagging it up front. Several open PRs authenticate Teams inbound image attachments (#47015, #55054) or touch Teams image handling (#48458). This PR covers the same authentication fix **and adds a piece none of them handle: a retry over the transient post-paste authorization window** (see below). Happy to fold this into an existing PR instead if a maintainer prefers — the retry is the part worth keeping regardless of which auth change lands.

## Problem

A pasted (inline) image in Microsoft Teams never reached the model — the agent replied as if it only received text, even with a multimodal model. File *uploads* worked fine.

## Root cause

A pasted inline image arrives as a Bot Framework attachment with `contentType image/*` and a **token-gated** `contentUrl` on `smba.trafficmanager.net` (`/v3/attachments/{id}/views/original`). The Teams adapter fetched it via `cache_image_from_url(content_url)` with **no `Authorization` header**, so the Connector returned **401**, the image was skipped, and `event.media_urls` stayed empty → `run.py`'s native image pipeline (gated on `media_urls`) was skipped → text only. File uploads use a pre-signed SharePoint `downloadUrl`, which is why they worked.

## Fix

- Fetch inline images with the bot's `api.botframework.com/.default` bearer and cache the bytes to a **local path** (mirrors the Discord/Telegram byte-cache path), so the native image pipeline attaches them.
- The client-credentials token mint is factored into a shared helper (`_mint_bot_connector_token`) reused by the standalone/cron send path; the adapter caches the token.
- **Transient window (the part unique to this PR):** a freshly-pasted attachment briefly returns 401/403 until the Connector authorizes the read — reproduced in production, where a `contentUrl` that returns 403 at the instant the message is delivered returns 200 a few seconds later (with the tenant-specific bearer; the `botframework.com`-issued token is rejected, i.e. the tenant-specific mint is the correct one). Without a retry the image is dropped permanently. The fetch retries with bounded backoff (`0/1/2/4/8s`) on an authenticated 401/403 plus the usual 408/429/5xx.
[...]

PR #97677, opened 2026-08-29
fix(teams): authenticate Bot Framework connector attachment downloads (pasted images 401)

kshitijk4poor:
## Summary

Pasted/inline images in Microsoft Teams now reach the agent instead of being silently dropped with a 401.

Teams delivers two kinds of inbound attachments: file uploads (paperclip) carry pre-authenticated SharePoint downloadUrls, while pasted/inline images carry a contentUrl on `smba.trafficmanager.net` (Bot Framework connector), which requires the bot's own bearer token. The adapter fetched all attachments unauthenticated, so every pasted image failed with `401 Unauthorized` and was dropped with only a log warning.

Root cause: no Bot Framework token acquisition for connector-hosted attachment downloads.

This is a consolidation of a five-way duplicate-PR cluster (#47015, #55054, #58476, #72977, #73685 all fix the same 401); the surviving implementation is based on #94547 with hardening folded in from review findings across the cluster.

## Changes

- `plugins/platforms/teams/adapter.py`:
  - `TeamsAdapter._get_botframework_token()` — client-credentials token for scope `https://api.botframework.com/.default`, cached until ~5 min before expiry (`time.monotonic()`), refresh serialized by an `asyncio.Lock` so concurrent attachments share one token POST (based on #94547 by @Sibbern; lock + monotonic clock from review of #94547/#55054)
  - `_fetch_attachment_bytes()` — attaches the bearer token only when the URL is an exact-match allowlisted Bot Framework host (`_ALLOWED_TEAMS_SERVICE_HOSTS`, https, default port); downloads stream through `_read_httpx_body_with_limit` so the shared inbound media cap applies (streaming + port from the sweeper review of #73685)
  - `_is_botframework_attachment_url()` — single predicate replacing the duplicated inline host checks. Exact-match allowlist, not suffix matching: any Azure customer can register `<name>.trafficmanager.net`, so `evil-trafficmanager.net` must not match. Also gates gov-cloud (`smba.infra.gov.teams.microsoft.us`)
  - Inbound `image/*` attachments with Bot Framework contentUrls route through the authenticated path + `cache_media_bytes()` instead of the unauthenticated `cache_image_from_url()`; images whose bytes fail validation are logged, not silently dropped; `media_types` records the computed MIME
  - Token-acquisition failure degrades to the pre-fix behavior (warning + unauthenticated fetch) — no new failure mode
- `tests/gateway/test_teams.py`: 8 new tests — exact-allowlist predicate (lookalike/subdomain/port/scheme negatives), BF routing vs generic helper, bearer attach + attacker-host block end-to-end, token acquire + cache reuse, concurrent cold-cache lock serialization, token-failure degradation, and the silent-drop regression guard

## Validation

| | Before | After |
|---|---|---|
| Pasted image (BF host) | 401 → dropped, only a log warning | cached + delivered to agent |
| Attacker host `evil-trafficmanager.net` / `emea.smba.…` / `:444` | n/a (would receive bearer in naive suffix match) | no token sent, no token acquired |
| Oversized / lying Content-Length | unbounded `.content` buffer | streamed, capped by shared media limit |
| Concurrent attachments, cold token cache | one STS POST per racing task | one shared POST (lock) |
| BF endpoint returns non-image bytes | silent drop, no log | warning logged |
| Token acquisition failure | warning + unauthenticated fetch (401) | same — no new failure mode |
| SharePoint paperclip uploads | work | unchanged (non-BF host → no token attached) |
| `tests/gateway/test_teams.py` | 24 passed | 32 passed (24 + 8 new) |
| Mutation checks | — | suffix match, silent drop, no-lock, unbounded buffer each fail a test |
| ruff / ty | — | clean / 0 new diagnostics |

E2E (real adapter code, isolated HERMES_HOME, mocked HTTP seam): 2 BF pasted images cached and delivered, token POSTed once and reused, `Authorization: Bearer` attached, file present on disk; lookalike/port/subdomain variants receive no bearer; 5 concurrent token fetches → 1 POST.

## Credit

Based on #94547 by @Sibbern (verified end-to-end on a live EMEA tenant by the author); his commit is cherry-picked with authorship preserved. Hardening folded in from the duplicate-PR cluster: exact-match allowlist approach of #73685 (@zenplace-system) and #55054 (@rodnavarro), `time.monotonic()` from #55054, streaming + port restriction from the automated review of #73685, token-refresh lock from the review of #94547.

Closes #94547

### DIFF
--- a/plugins/platforms/teams/adapter.py
+++ b/plugins/platforms/teams/adapter.py
@@ -898,29 +898,75 @@ class TeamsAdapter(BasePlatformAdapter):
         self._mark_disconnected()
         logger.info("[teams] Disconnected")
 
+    async def _get_botframework_token(self) -> str:
+        """Acquire a Bot Framework bearer token via client credentials.
+
+        Needed to download connector attachments (smba.trafficmanager.net
+        /v3/attachments/...), which -- unlike SharePoint file downloadUrls --
+        are NOT pre-authenticated and return 401 without the bot's own
+        token. Token is cached until ~5 minutes before expiry.
+        """
+        import time
+        import httpx
+
+        cached = getattr(self, "_bf_token_cache", None)
+        if cached and cached[1] > time.time() + 300:
+            return cached[0]
+
+        client_id = self._client_id
+        client_secret = self._client_secret
+        tenant_id = self._tenant_id
+        if not (client_id and client_secret and tenant_id):
+            raise ValueError("Missing TEAMS_CLIENT_ID/SECRET/TENANT_ID for attachment auth")
+
+        async with httpx.AsyncClient(timeout=15.0) as client:
+            resp = await client.post(
+                f"https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token",
+                data={
+                    "grant_type": "client_credentials",
+                    "client_id": client_id,
+                    "client_secret": client_secret,
+                    "scope": "https://api.botframework.com/.default",
+                },
+            )
+            resp.raise_for_status()
+            payload = resp.json()
+        token = payload["access_token"]
+        self._bf_token_cache = (token, time.time() + int(payload.get("expires_in", 3600)))
+        return token
+
     async def _fetch_attachment_bytes(self, url: str, timeout: float = 30.0) -> bytes:
         """Download attachment bytes with SSRF protection.
 
         Teams file attachments carry pre-authenticated SharePoint download
-        URLs (no extra auth header needed). Validates the URL against the
-        SSRF guard and follows redirects through the shared redirect guard,
+        URLs (no extra auth header needed). Bot Framework connector
+        attachment URLs (pasted/inline images on smba.trafficmanager.net /
+        botframework.com hosts) require the bot's bearer token -- detected
+        below and fetched with auth. Validates the URL against the SSRF
+        guard and follows redirects through the shared redirect guard,
         matching the cache_*_from_url helpers in gateway.platforms.base.
         """
+        from urllib.parse import urlparse
         from tools.url_safety import create_ssrf_safe_async_client, is_safe_url
         from gateway.platforms.base import _ssrf_redirect_guard
 
         if not is_safe_url(url):
             raise ValueError("Blocked unsafe attachment URL (SSRF protection)")
 
+        headers = {"User-Agent": "Mozilla/5.0 (compatible; HermesAgent/1.0)"}
+        host = (urlparse(url).hostname or "").lower()
+        if host.endswith("trafficmanager.net") or host.endswith("botframework.com"):
+            try:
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed `plugins/platforms/teams/adapter.py` + tests for PR #97677. The fix is well-targeted: pasted/inline images arrive on Bot Framework connector hosts that need the bot's own bearer token, and the change routes them through an authenticated fetch while leaving SharePoint `downloadUrl` file uploads untouched. The dot-anchored `_is_botframework_attachment_host` predicate is correct (rejects `evil-trafficmanager.net` / `notbotframework.com`, verified by tests), the token is cached with a 5-minute pre-expiry buffer, and token-acquisition failure degrades to a warned unauthenticated fetch rather than crashing. Test coverage is strong, including the attacker-lookalike case and the invalid-bytes warning regression guard.

Non-blocking observations:

1. **Bearer token and cross-host redirects.** The `Authorization` header is attached to the original request and the client follows redirects (`follow_redirects=True`) with `_ssrf_redirect_guard` on the response event. httpx strips `Authorization` on cross-host redirects by default, and the guard validates redirect targets, so the token shouldn't leak — but this is the one token-exfiltration boundary in the change. Worth confirming (ideally with a test) that a BF-host URL that 302s to an attacker host is both blocked by `_ssrf_redirect_guard` and never carries the `Authorization` header on the cross-host request, so the safety doesn't silently depend on httpx internals.

2. **Gov/sovereign cloud hosts are not covered.** The adapter already whitelists government endpoints (`smba.infra.gov.teams.microsoft.us` in `_ALLOWED_TEAMS_SERVICE_HOSTS`), but `_is_botframework_attachment_host` only matches `*.trafficmanager.net` / `*.botframework.com`, and the token endpoint is hardcoded to `login.microsoftonline.com`. Pasted images in gov/sovereign-cloud tenants will therefore still 401 (no token attached), and even if they were detected, the token request would go to the wrong authority (`login.microsoftonline.us` for GCC High/DoD). If gov-cloud support is in scope for Teams, the host predicate and token authority need to be tiered; if it's explicitly out of scope, a comment documenting that would prevent future confusion.

3. **No retry on transient token-endpoint failure.** `_get_botframework_token` performs one POST with a 15s timeout; a transient network blip fails the whole attachment with a warning and an unauthenticated fetch that will 401 anyway. A single bounded retry (or a short-lived negative cache) would make pasted images more resilient without much complexity. Minor.

4. **Minor:** `_get_botframework_token` mutates `self._bf_token_cache` without a lock — fine for the adapter's serialized loop, but worth a note if the adapter ever processes messages concurrently. Also, `_fetch_attachment_bytes` does a fresh `login.microsoftonline.com` round-trip per adapter instance on cold cache; acceptable.

The change matches its stated symptom and is well tested; nothing blocking.

PR #97683, opened 2026-08-29
fix(desktop): annotate canonical relay targets for remote @mentions

### LINKED ISSUE #97678: [Bug]: Desktop remote @mention emits an unresolvable message_agent target from its UI alias
## Bug description

Desktop Bot Mode uses a UI-facing multi-source mention alias as if it were a canonical `message_agent` relay identity.

For a connected-machine bot, the mention middleware currently computes:

```ts
const handle = botHandle(bot.name, bot)
const target = bot.remoteSource && bot.connectionId ? `${handle}@${bot.connectionId}` : handle
```

`botHandle()` prefers the roster row's precomputed UI handle, which can be a source-qualified alias such as `default-remote-1`. The middleware then instructs the active agent to call:

```text
message_agent target: "default-remote-1@remote-1"
```

The relay resolver does not recognize that UI alias. It accepts only the registered relay `handle` or `profile`, optionally qualified by the connection ID. For a roster row whose canonical values are `handle=hermes`, `profile=default`, and `connection_id=remote-1`, valid targets include `hermes@remote-1` and `default@remote-1`; `default-remote-1@remote-1` cannot resolve.

The agent therefore follows Desktop's explicit hidden instruction and receives `No teammate named ...` before the message is enqueued.

## Steps to reproduce

1. Connect Desktop to a second Hermes runtime and expose the same profile name across more than one source so the union roster creates a source-qualified UI mention alias.
2. From a Bot Chat with `message_agent` available, select the remote bot from Desktop's `@mention` autocomplete and ask the active agent to contact it.
3. Inspect the hidden mention annotation added to the submitted user message.
4. Observe that Desktop supplies `message_agent target: "<ui-alias>@<connection-id>"`.
5. Observe that `message_agent` rejects the target because neither the relay handle nor profile equals the UI alias.
6. Retry manually with `<canonical-handle>@<connection-id>` or `<profile>@<connection-id>`; the relay accepts the target.

## Expected behavior

The target emitted by Desktop's mention middleware must be resolvable against the synchronized relay roster. UI aliases and transport identities may differ, but the middleware should emit the canonical relay handle/profile or carry a structured identity that the relay resolves deterministically.

## Actual behavior

Desktop emits a source-qualified UI alias and then appends the connection ID a second time. The resulting target is outside the relay resolver's documented contract and fails before enqueue.

## Impact

- Remote bot handoffs selected through Desktop autocomplete can fail immediately.
- Recovery depends on the model independently inferring a different canonical identity.
- The failure can be mistaken for an offline bot or for the separate long-running relay timeout in #93911.

## Root cause and coverage gap

Current `origin/main` (`9f90cd438c`) still contains the mismatch:

- `apps/desktop/src/plugins/hermes-bots/plugin.tsx`: the mention middleware builds the target from `botHandle(...)` plus `@connectionId`.
- `apps/desktop/src/plugins/hermes-bots/data.ts`: `botHandle(...)` prefers the UI row's precomputed `bot.handle`.
- `tools/bot_relay.py`: `resolve_remote_target(...)` matches only `row.handle` or `row.profile`, optionally against `row.connection_id`.

The existing mention test asserts that a remote target annotation exists, but it pins an alias-qualified form without checking whether the target resolves against a real relay roster. A cross-layer regression test should feed the emitted target into `resolve_remote_target()`.

## Environment and evidence

- Windows 11 Hermes Desktop connected to a WSL2 Hermes runtime.
- Incident checkout: `b3a2065ff3`.
- Rechecked against current `origin/main`: `9f90cd438c`.
- Persisted incident sequence, sanitized:
  1. Desktop annotation supplied an alias-qualified target.
  2. The agent used it verbatim.
  3. The relay rejected it immediately as unknown.
  4. A retry with the canonical registered identity was accepted.

No private paths, host/profile labels, request IDs, or conversation content are included here.

## Related but distinct

- #93911: a valid relay request is accepted, then Desktop abandons `bot_relay.deliver` after 30 seconds.
- #94018: ordinary Desktop sessions can lack the structured `message_agent` capability.
- #91397: a resolved routing mention was forwarded verbatim to the recipient.

None of those reports covers Desktop generating an unresolvable `message_agent` target from its own UI alias.

liuhao1024:
## What does this PR do?

Desktop Bot Mode's mention middleware annotated `message_agent` targets from `botHandle()`, which prefers a roster row's **source-qualified UI alias** (`default-vera`). Neither resolver accepts that form: the relay's `resolve_remote_target()` matches only a canonical roster `handle`/`profile` (optionally `@connection-id`-qualified), and the local path resolves only a bare profile name or `hermes`. A remote handoff selected through `@mention` autocomplete therefore died with `No teammate named ...` before the message was ever enqueued, even though the agent followed Desktop's own hidden instruction verbatim.

The middleware now annotates the canonical identity instead: `profile@connection-id` for remote rows, and the canonical bare handle (`default` → `hermes`) for local rows. The UI alias still appears in the human-readable identification line — only the machine-resolved target changes. The relay side gains a contract pin for the `profile@connection-id` form, so the emitted target stays inside the documented resolver contract in both directions.

## Related Issue

Fixes #97678

## Type of Change

## Changes Made

- `apps/desktop/src/plugins/hermes-bots/plugin.tsx` — the mention middleware builds its `message_agent target:` annotation from the canonical identity (`${bot.name}@${bot.connectionId}` for remote rows, `botHandle(bot.name)` for local rows) instead of the UI alias `botHandle(bot.name, bot)`.
- `apps/desktop/src/plugins/hermes-bots/plugin.mentions.test.ts` — the target assertion now pins the canonical `default@vera` form and asserts the alias-qualified form is gone.
- `tests/tools/test_bot_relay.py` — `test_resolve_remote_target_forms` pins `resolve_remote_target("default@cloud-1", ...)` so the `profile@connection-id` form the middleware emits stays resolvable.

## How to Test

1. `cd apps/desktop && npx vitest run src/plugins/hermes-bots/plugin.mentions.test.ts` — Observed result: 15 passed, including `hands the agent a relay-resolvable canonical target` (`message_agent target: "default@vera"`).
2. `npx vitest run src/contrib/plugin.test.ts` (the other suite importing the plugin) — Observed result: 6 passed.
3. `npx tsc -p . --noEmit` — Observed result: exit 0.
4. `pytest tests/tools/test_bot_relay.py -q` — Observed result: 26 passed, including the new `default@cloud-1` pin.

### Code

- [x] I've tested on my platform: macOS 26.0 (arm64)

## Screenshots / Logs

Not applicable — the change is inside a hidden annotation string; the visible `@handle` identification line is unchanged.

### DIFF
--- a/apps/desktop/src/plugins/hermes-bots/plugin.mentions.test.ts
+++ b/apps/desktop/src/plugins/hermes-bots/plugin.mentions.test.ts
@@ -253,11 +253,15 @@ describe('the mention middleware', () => {
     expect(hostMock.requestProfile).not.toHaveBeenCalled()
   })
 
-  it('hands the agent the connection-qualified message_agent target', async () => {
+  it('hands the agent a relay-resolvable canonical target', async () => {
     const { handler } = await contributions()
     const result = await handler({ text: 'ping @default-vera' })
 
-    expect(result.text).toMatch(/message_agent target: "default-vera@vera"/)
+    // The UI alias ('default-vera') is not a relay identity: resolve_remote_target()
+    // accepts only a roster row's handle/profile, optionally @connection-qualified.
+    // The annotation must carry the canonical profile@connection form (#97678).
+    expect(result.text).toMatch(/message_agent target: "default@vera"/)
+    expect(result.text).not.toMatch(/message_agent target: "default-vera/)
     expect(result.text).toMatch(/on Vera/)
   })
 
--- a/apps/desktop/src/plugins/hermes-bots/plugin.tsx
+++ b/apps/desktop/src/plugins/hermes-bots/plugin.tsx
@@ -730,7 +730,13 @@ export default {
               botRosterMeta(bot, $botMeta.get())?.title || bot.ui_meta?.['hermes-bots']?.title || bot.title || ''
             ).trim()
 
-            const target = bot.remoteSource && bot.connectionId ? `${handle}@${bot.connectionId}` : handle
+            // message_agent only resolves canonical identities: the relay
+            // matches a roster row's handle/profile (± @connection-id), the
+            // local path a bare profile name or 'hermes'. botHandle() prefers
+            // the row's source-qualified UI alias ('default-vera'), which
+            // neither resolver accepts — annotate the canonical form instead.
+            const target =
+              bot.remoteSource && bot.connectionId ? `${bot.name}@${bot.connectionId}` : botHandle(bot.name)
 
             const where = bot.remoteSource
               ? ` — on ${bot.connectionLabel || bot.connectionId} (message_agent target: "${target}")`
--- a/tests/tools/test_bot_relay.py
+++ b/tests/tools/test_bot_relay.py
@@ -84,6 +84,9 @@ def test_resolve_remote_target_forms(root):
     assert bot_relay.resolve_remote_target("default", roster)["connection_id"] == "cloud-1"
     # exact connection-qualified form
     assert bot_relay.resolve_remote_target("hermes@cloud-1", roster)["profile"] == "default"
+    # profile@connection — the form Desktop's mention middleware annotates
+    # for remote bots (#97678); the UI alias form must not be required
+    assert bot_relay.resolve_remote_target("default@cloud-1", roster)["profile"] == "default"
     assert bot_relay.resolve_remote_target("hermes@nope", roster) is None
     assert bot_relay.resolve_remote_target("ghost", roster) is None
 


Enough1122:

Reviewed `apps/desktop/src/plugins/hermes-bots/plugin.tsx` + tests for PR #97683. The root cause is clearly identified and the fix is minimal: `message_agent` targets were annotated from `botHandle()`'s source-qualified UI alias (`default-vera`), which neither `resolve_remote_target()` (canonical handle/profile ± `@connection-id`) nor the local path accepts. Switching the remote branch to `${bot.name}@${bot.connectionId}` and adding a relay test for the `profile@connection` form is the right, narrow fix. The test updates accurately pin the new canonical form and guard against the alias leaking back in.

Non-blocking observations:

1. **The local branch still passes through `botHandle(bot.name)`.** For a remote bot the alias is now avoided, but a non-remote (local) bot still gets `target = botHandle(bot.name)`. If `botHandle()` ever returns the source-qualified UI alias for a local row too (the comment describes it as preferring that alias generally), the local `message_agent` target would hit the same "not resolvable" bug the PR fixes for remote bots. A test covering the local (no `remoteSource`) path — asserting the target is the bare canonical profile name — would make the invariant explicit. (If `botHandle` is known to return the canonical name for local rows, a brief comment would suffice.)

2. **Minor:** `bot.name` is now load-bearing for the remote target; an empty/undefined `bot.name` would produce a malformed `@connection` target. Likely impossible for roster rows, but a guard (or a fallback to `handle`) would be cheap.

The change is consistent with the relay's documented resolution rules and the regression tests are well placed; nothing blocking.

liuhao1024:
Thanks for the review. Both observations are fair. I deliberately kept the change scoped to the remote-canonical form: the local branch keeps the roster's taggable handle, which is the same form the mention middleware produces for local rows. An explicit local-path assertion (and an empty-name guard) would be reasonable hardening — happy to fold either in if maintainers want them.

teknium1:
Merged via #101942 (rebase-merge, your commit 2e542c92e67 kept with authorship — thanks @liuhao1024). Follow-up on top widened the same annotation to the LOCAL twin (`default-this-device` when a remote gateway is active), which the Discord reporter's video showed. Fixes #97678.

PR #95918, opened 2026-08-26
feat(gateway): add governed generic specialist handoffs

mrkillbob:
Replaces the specialist-orchestration and Kanban metadata portion of #94080 as a self-contained generic change. Routing is explicit opt-in, bounded, tool-free, fail-closed, configurable, neutral, and refuses profiles that are not installed. Handoffs are durable and idempotent; goal/skill contracts propagate through decomposition; subscribers receive decomposition and truthful circuit-breaker notices; bound-board metadata retains project identity.

Verification: 31 focused tests passed after the final update; targeted Ruff passed; git diff --check passed. Rebased onto current main. Part of #94080 split.

PR #97684, opened 2026-08-29
feat(kanban): govern specialist discovery approvals

mrkillbob:
## Summary
- add fail-closed specialist candidate, benchmark, verification, canary, revocation, and recovery lifecycle controls
- route specialist handoffs through the board-local active capability registry and preserve task-orchestrator fallback
- add authenticated dashboard operator approval using an explicit OAuth subject allowlist; loopback session tokens cannot approve

## Verification
- focused specialist/approval suite: 67 passed
- ruff and git diff --check passed
- full suite bounded check is blocked by unrelated tests/agent/lsp/test_client_e2e.py::test_client_lifecycle_clean: the test live-system guard rejects terminating an LSP subprocess after 156 tests passed

No provider/model/network call, live trading action, or generated-profile activation was performed.


mrkillbob:
Superseded as a cumulative specialist-discovery train. The focused replacement is #97870, based on the specialist-routing work in #95918. The source branch is retained; no commits were deleted or force-pushed.

PR #89613, opened 2026-08-19
fix(plugins): stop the security scanner from reading test trees

liuhao1024:
## What does this PR do?

`plugin_guard` scans the whole plugin clone (`plugin_dir.rglob("*")`), and `EXCLUDED_DIRS` skipped caches and vendored dirs — but not `tests/`. A security-conscious plugin's test suite **should** contain adversarial fixtures: a test asserting the trust boundary holds round-trips the injection string verbatim, and the scanner reads it as the attack it is testing for. Any single `critical` finding makes the verdict `dangerous`, which `--force` explicitly cannot override — so scanning tests made exactly the plugins that test their security **unconditionally uninstallable** (#89610, reported downstream as yantrikos/yantrikdb-hermes-plugin#67: a plugin with 3 criticals, 2 of them in test files, went from `dangerous` (cannot install) to `caution` (installs with confirmation) purely by excluding the test tree).

The only workaround was obfuscating the payload strings (concatenating fragments so the scanner misses them) — which weakens real tests, teaches authors to hide strings from the scanner, and inverts the incentive: plugins that test their trust boundary get punished relative to plugins that don't test at all. Test fixtures are never loaded into an agent's context at runtime the way `README` or `plugin.yaml` are — the same distinction `CODE_EXEMPT_PATTERN_IDS` already draws for code files.

This PR adds the conventional test-tree names (`tests`, `test`, `testing`, `spec`, `specs`, `fixtures`) to `EXCLUDED_DIRS`, alongside the existing cache/vendored skips. Production-code findings — including every critical in shipped code — are untouched and still fail closed exactly as before.

## Related Issue

Fixes #89610

## Type of Change

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `tools/plugin_guard.py` — `EXCLUDED_DIRS` gains the test/spec/fixture directory names, with a comment stating why fixtures are categorically different from shipped docs/config (never loaded into an agent's context; adversarial content is their purpose).
[...]

PR #97686, opened 2026-08-29
fix(plugin-guard): exclude tests/ from plugin scan (closes #93927)

### LINKED ISSUE #93927: plugin_guard: destructive strings in TEST fixtures produce DANGEROUS verdict, blocking install of legitimate plugins with no override
## Summary

`tools/plugin_guard.py` scans **all** files in a plugin tree, including `tests/`. Test fixtures that deliberately exercise hostile shell input (e.g. a destructive-remove string used as a quoting edge case) are scored as `critical/destructive` runtime threats. Per `skills_guard._determine_verdict` (line 1153), a single `critical` finding → verdict `dangerous`, and `dangerous` is always blocked (`should_allow_plugin_install`) — `--force` cannot override.

**Net effect: any plugin repo that ships security-focused unit tests cannot be installed at all.**

## Minimal repro (scanner pinned at plugin-guard-v1)

```
plugin.yaml          # name/description/manifest_version: 1
tests/test_shell.py  # contains: tricky = "do not <destructive root rm> $HOME `tick`"
```

```python
from pathlib import Path
from tools.plugin_guard import scan_plugin
r = scan_plugin(Path('guard-repro'), source='repro')
# verdict: dangerous
# - critical destructive tests/test_shell.py:2 destructive_root_rm
# - critical destructive tests/test_shell.py:2 destructive_home_rm
```

Real-world case: installing `jmanhype/paivot-hermes` yields **DANGEROUS, 110 findings** — all from its own test suite (e.g. `tests/test_driver_ops.py:236`, a shell-quoting robustness test; `tests/test_vault_s5_evolve.py:226`, a subprocess test). The plugin's runtime code is clean; only fixtures trip it. Workaround used: `plugins.scan_on_install: false` for the install window, then re-enable — which silently skips the whole scan rather than just the tests.

(Note: even filing this issue through the Hermes agent required routing around its terminal hardline block, because the issue text itself contains the pattern strings — same false-positive class, different layer.)

## Why tests/ is different

Test code is never imported at plugin runtime; destructive *strings* inside fixtures assert the plugin handles hostile input safely. Scoring them at the same severity as runtime destructive commands conflates the vaccine with the disease.

## Suggested directions

1. Exclude `tests/` (or files named `test_*.py` / `*_test.py`) from the destructive/exec pattern tiers in `plugin_guard.scan_plugin` — `EXCLUDED_DIRS` (line 62) is the existing mechanism, but full exclusion may be too broad for e.g. conftest network fixtures; a severity downgrade for test files may be safer.
2. A per-plugin trust/allowlist marker for self-authored sources (config-declared), since `trust_level` is hardcoded `community` in `scan_plugin`.
3. At minimum: allow an explicit operator override for `dangerous` on first-party installs (config-gated), instead of no path at all.

## Environment

- Hermes CLI at `~/.hermes/hermes-agent` (`PLUGIN_SCANNER_VERSION = "plugin-guard-v1"`)
- macOS 15.6.1, Python 3.12

jmanhype:
Fixes #93927.

## Problem
scan_plugin rates plugin repos DANGEROUS (always-blocked) when their test suites contain scanner-bait fixtures — destructive-looking strings, decoy secrets, or system-file paths that exist precisely to prove the plugin's own defenses REJECT them.

Live repro from the issue: paivot-hermes scored DANGEROUS (126 findings, 4 critical) — all 4 criticals inside tests/ (shell-quoting regression fixtures, decoy-token secret-scrub test, traversal-rejection tests). None executed at runtime.

## Fix
Add `tests` to `EXCLUDED_DIRS` in `tools/plugin_guard.py`. Test code never ships in the runtime path; fixtures exercise defensive rejection, not destructive behavior.

## Verification (live, post-fix)
```
verdict=caution  findings=34  severities={'medium': 27, 'high': 7}
```
Exactly matches the minus-tests/ projection from the issue triage (0 critical, 7 high). Remaining highs are vendored gitignored README install lines and negative doc statements — caution is the correct human-confirmable posture for self-authored source.

### DIFF
--- a/tools/plugin_guard.py
+++ b/tools/plugin_guard.py
@@ -59,9 +59,14 @@ from tools.skills_guard import (
 PLUGIN_SCANNER_VERSION = "plugin-guard-v1"
 
 # Directories that are never scanned (VCS internals, caches, vendored envs).
+# "tests" is excluded because test fixtures deliberately contain destructive
+# strings (rm -rf /, secret decoys, /etc/passwd) to prove the code under test
+# REJECTS them — scoring them as production threats yields DANGEROUS verdicts
+# for self-authored plugin repos (see hermes-agent#93927).
 EXCLUDED_DIRS = {
     ".git", "__pycache__", "node_modules", ".venv", "venv",
     ".mypy_cache", ".pytest_cache", ".ruff_cache", ".tox",
+    "tests",
 }
 
 # Code file extensions where "reads an env secret" / "HTTP call with a key


alt-glitch:

Duplicate of #89613: both exclude plugin test/fixture trees from plugin_guard scanning to prevent fixture strings from forcing a DANGEROUS verdict.

teknium1:
Your `tests/` exclusion is the same fix as #89613 (filed earlier, wider dir set); both are landing via #110218 with credit to you in the body. This PR will be closed with credit once that merges. Thank you!

teknium1:
Thanks @jmanhype — this landed on main via #110218 (`1e76efbe28ef`) with your commits/credit preserved. Final shape: test trees are still scanned; a critical finding under a root-level test dir caps the verdict at caution (confirm / --force) while a critical in runtime code still yields dangerous.
Closing in favour of the merged salvage.

PR #97688, opened 2026-08-29
fix(config): register file_tools.hosted_ocr, opt-in keyless hosted OCR

### LINKED ISSUE #97687: [Bug]: file_tools.hosted_ocr is not a registered config key so the hosted OCR switch is undiscoverable
### Bug Description

#97195 added a `file_tools.hosted_ocr` config switch for hosted PDF OCR in `tools/read_extract.py`. The key was never registered in `DEFAULT_CONFIG`, so the config system does not know it exists.

This key is the only off switch for an automatic outbound call (the document is sent to Firecrawl's Parse endpoint). Right now there is no supported way to discover it:

- `hermes config set file_tools.hosted_ocr false` saves the value but prints an unknown-key notice and suggests `tools.hosted_ocr`, a key that does nothing.
- `cli-config.yaml.example` does not mention it.
- The setup wizard cannot prompt for it.

There is also a free capability locked out by accident. The bundled `firecrawl-anydoc` 0.2.4 supports keyless hosted OCR by design. Its docstring says the api key "falls back to `FIRECRAWL_API_KEY`, then keyless". The gate from #97195 only runs hosted OCR when a key is set, so users without a Firecrawl account get the NEEDS OCR warning even though the bundled converter would extract the text for free.

I verified the keyless route live with no key set. An image-only PDF (confirmed to raise `NeedsOcrError` on the default call) converted through `ocr='hosted'` three times in a row, each in under 2.5 seconds, with correct text back from `api.firecrawl.dev`.

### Steps to Reproduce

1. Run `hermes config set file_tools.hosted_ocr false` on current main.
2. Notice the unknown-key warning and the wrong suggestion.
3. Search `cli-config.yaml.example` for `hosted_ocr`. Nothing.
4. Read a scanned PDF with no `FIRECRAWL_API_KEY` set. You get the NEEDS OCR warning even though the keyless free tier is one call away.

### Expected Behavior

- `file_tools.hosted_ocr` is a recognized key, documented in the example config, with a stable default.
- The default stays key-gated. Setting `FIRECRAWL_API_KEY` is the consent signal for sending a document to Firecrawl, that decision from #97195 should not change.
- There is an explicit opt in for users without a key who want the free tier.

### Actual Behavior

- The key is read by `tools/read_extract.py` but absent from `DEFAULT_CONFIG`, so config tooling treats it as unknown.
- No opt in exists for keyless, the library's keyless fallback is unreachable through Hermes.

### Affected Component

- Configuration (config.yaml, .env, hermes setup)

N/A (CLI only)

### Debug Report

```shell
N/A. Found by reading #97195 against current main. The unknown-key notice and the keyless route were both verified locally with unit tests and a live conversion probe.
```

### Operating System

Windows 11

### Python Version

3.11

### Hermes Version

main 9f90cd438

`tools/read_extract.py` reads `file_tools.hosted_ocr` through `load_config_readonly()`, but `hermes_cli/config_defaults.py` never gained the `file_tools` section, so `_validate_config_key` classifies it as unknown. Per AGENTS.md, new config keys go into `DEFAULT_CONFIG` (a new key in a new section needs no version bump, the deep merge covers it).

### Proposed Fix (optional)

Register `file_tools` with `hosted_ocr: null` in `DEFAULT_CONFIG` and document it in `cli-config.yaml.example`. Keep null (default) meaning key-gated exactly as shipped. Add `true` as an explicit opt in that also allows the keyless free tier, and keep `false` as the hard off. This keeps the #97195 consent-by-key default untouched while unlocking the free route for users who ask for it.

Adolanium:
Allow an explicit file_tools.hosted_ocr setting to opt into keyless hosted document OCR. True enables it, false disables uploads even with a key, and null retains the existing key-gated default. Hand-edited boolean strings follow the same policy.

Register the setting in config defaults and the existing agent settings category. Preserve profile-scoped credential resolution and pass an explicit empty key for keyless requests so anydoc cannot fall back to another profile's process credentials or endpoint.

Fixes #97687

Validation: 148 targeted tests passed through scripts/run_tests.sh across config and read_file OCR gating. Tests read real temporary YAML into runtime/schema resolution and capture the real anydoc SDK request under a keyless multiplexed profile, verifying no inherited Authorization header and the intended endpoint.

Updated on the refactored main at 1a990f3062, with the repair folded into one commit.


Enough1122:

Reviewed `tools/read_extract.py`, `hermes_cli/config_defaults.py`, `cli-config.yaml.example` + tests for PR #97688. Commit 1 (registering `file_tools.hosted_ocr` in `DEFAULT_CONFIG` + the example config) fixes a real gap from #97195 — an unregistered key silently ignored by the deep-merge/setup. Commit 2's resolution logic (false = off, true = keyless opt-in, null/default = key-gated) is coherent, well-documented, and the refactor of `hosted_ocr_available()` to delegate to `_hosted_ocr_config()` removes genuine duplication. Tests cover the gate states and the keyless retry path (`ocr='hosted'` with no `api_key`).

Non-blocking observations:

1. **`is True` / `is False` identity checks ignore non-bool values.** `setting is True` / `setting is False` — a YAML value that parses to something else (e.g. a quoted `"true"` string, `1`, `yes`) would be silently treated as "null/auto" and key-gated. The prior code used the same `is False` convention so this is consistent, but `== True`/`== False` (or a small normalization) would be more forgiving of config typos.

2. **Schema wording vs. keyless mode.** `hosted_ocr_available()` now returns True for `hosted_ocr: true` with no key, which drives the `read_file` schema line ("PDF (scanned or text)"). Worth confirming the schema text is mode-agnostic and doesn't tell a keyless opt-in user they must set `FIRECRAWL_API_KEY` (the NEEDS-OCR warning text is referenced as the fallback curriculum, but the schema line is what the model reads at call-build time).

3. **Keyless egress is an upload with no account/credential.** With `hosted_ocr: true` and no key, scanned PDFs go to firecrawl.dev's free tier unauthenticated — rate limits and any data-retention caveats of the keyless tier are now the user's exposure. The config comment documents this ("sends the document to Firecrawl's Parse endpoint"), which is the right consent framing; just flagging it so the trade-off is consciously shipped.

4. **Minor:** the `test_runtime_keyless_route_with_opt_in` patch chains use inline-whitespace line breaks (e.g. `patch.object(rx, "_hosted_ocr_config",` newline `return_value=...`) that read awkwardly; cosmetic only.

Both commits are internally consistent and the registration fix stands alone as claimed; nothing blocking.

Adolanium:
Thanks for the review. Went through all four points:

1. Valid, fixed in e4c7a2f88 (folded into the feature commit so it is correct as a unit). Quoted values from hand-edited YAML now go through a tri-state coercion using the same token table as `hermes config set` (true/1/yes/on, false/0/no/off). The expensive direction was a quoted `'false'` silently failing to disable the outbound call when a key is set. Anything unrecognized still reads as auto, matching the repo's existing `Optional[bool]` coercers (same shape as `tools/video_generation_tool.py`). New test covers quoted true/false, `yes`, quoted `1`, and unrecognized values (string, int, list).

2. Verified moot: the schema line is exactly one dynamic word, `PDF (text layer)` becomes `PDF (scanned or text)`. Nothing in the schema text mentions `FIRECRAWL_API_KEY` or any route, for either mode.

3. Agreed on the framing, no code change. The default (null) stays key-gated and the config comment states the upload in plain words, so keyless only happens for a user who wrote `true` themselves.

4. Left as-is: the wide whitespace runs in the test file match the style of the surrounding `#97195` tests, so keeping them consistent reads better than reflowing just the new lines.

Adolanium:
Follow-up found in the same area: the bytes door of anydoc extraction (used when read_file transfers a document from a remote backend) had no NeedsOcrError handling, so the same scanned PDF that gets OCR or the NEEDS-OCR warning locally produced a bare convert error over SSH. Fixed in #97721 (issue #97720), reusing the same gate and warning this PR registers.

Adolanium:
I updated this on the refactored main and pushed e71b7c7ae2 as one commit. I fixed the settings-category integration and rebased onto the refactored code. I also fixed the SDK fallback that could make a keyless profile use the process-default Firecrawl credential, with a regression test against real SDK request construction.

148 targeted tests passed through scripts/run_tests.sh across config and read_file OCR gating. Tests read real temporary YAML into runtime/schema resolution and capture the real anydoc SDK request under a keyless multiplexed profile, verifying no inherited Authorization header and the intended endpoint.

Adolanium:
I resolved the new upstream test conflict and rebased this onto 1a990f3062, preserving both the current config validation assertions and the hosted OCR test. Pushed 5d2050a5b1 as one commit. All 148 config/OCR tests passed through scripts/run_tests.sh after the rebase.

PR #97689, opened 2026-08-29
feat(delegation): subagents can see images — per-task image forwarding (port of Roomote#1796)

teknium1:
## Summary
`delegate_task` tasks can now carry an `images` list — subagents finally SEE the screenshot/mock/chart the task is about, instead of working blind from a text description.

Port of RooCodeInc/Roomote#1796 / #1767 (Fast agent forwards bounded current-turn attachments to delegated coding tasks), adapted to Hermes' delegation boundary.

## Changes
- `tools/delegate_tool.py`: per-task `images` (max 8; local paths or http(s) URLs), validated loudly before any child spawns; forwarded on the child's goal turn. Vision-capable children get native `image_url` content parts (local files as data URLs, remote URLs verbatim); non-vision children get `[Image attached at: …]` hints + a `vision_analyze` pointer. Routing reuses `agent.image_routing` (`decide_image_input_mode` / `build_native_content_parts`) so `agent.image_input_mode` governs delegation exactly like inbound gateway images. Schema advertises the per-task field; legacy top-level single-goal shape also accepts `images`.
- `run_agent.py`: `_dispatch_delegate_task` forwards `images` (single dispatch site, all invocation paths).
- `tests/tools/test_delegate_images.py`: 17 new tests (validation, native/text/degradation paths, schema surface).
- `website/docs/user-guide/features/delegation.md`: "Forwarding Images to a Subagent" section.

Fail-safe by contract: unreadable paths skipped with a log line; any exception in the image plumbing degrades to the plain text goal — image forwarding can never break a spawn. Message-alternation and prompt-cache invariants untouched (the multimodal content rides the child's normal first user turn; parent context sees only the delegation call).

## Validation
| | Result |
|---|---|
| `tests/tools/test_delegate_images.py` (new) | 17/17 pass |
| `test_delegate.py` + `test_delegate_batch_validation.py` + `test_delegate_output_schema.py` | 123/123 pass |
| E2E (real imports, temp HERMES_HOME) | malformed `images` rejected pre-spawn; real PNG → `['text','image_url','image_url']` parts |

**Live repro:** n/a (new capability, not a bugfix); behavior exercised E2E with real function calls and a real PNG on disk.

Source: https://github.com/RooCodeInc/Roomote/pull/1796 and https://github.com/RooCodeInc/Roomote/pull/1767

## Infographic

Subagents can now see images


tools/delegate_tool.py
```diff
@@ -2451,6 +2451,128 @@ def _apply_summary_budget(results: List[Dict[str, Any]], parent_agent) -> None:
...
+        except Exception:
+            cfg = None
+
+        mode = decide_image_input_mode(
+            str(getattr(child, "provider", "") or ""),
+            str(getattr(child, "model", "") or ""),
+            cfg,
+        )
+        if mode == "native":
+            parts, skipped = build_native_content_parts(goal, paths, urls)
```
unsupportedpastels, line 2537: Could we bind these paths to actual current-turn attachments rather than accepting arbitrary model-generated filesystem paths? As written, `images` can name any readable local file, and `build_native_content_parts()` will base64-encode it for the child provider. Unrecognized content is currently labeled `image/jpeg`, so I confirmed that a non-image `.txt` file is forwarded byte-for-byte. This can also provide a file-read/exfiltration path when the parent has delegation available but not file-reading tools. One safe shape would be opaque attachment IDs resolved through a session-bound attachment manifest; if general paths are intentional, it would help to enforce equivalent read authority, approved roots, and successful image magic/decode validation before sending any bytes.

PR #97692, opened 2026-08-29
fix(skills): roll back a batch when an op raises instead of returning an error

### LINKED ISSUE #97691: [Bug]: a raised op in a skill_manage batch skips rollback and leaves the batch half applied
### Bug Description

The `operations[]` batch in `skill_manage` (#97295) is atomic for ops that return an error result: the batch rolls every touched skill back. But an op that RAISES instead of returning error JSON (an I/O error, a permission error, any unexpected bug) propagates out of the execution loop.

The `finally` block then deletes the snapshot directories before any rollback can run. Result: earlier ops stay applied, the recovery snapshots are gone, and the model only sees the generic tool error from the outer wrapper. The batch is left half applied with no recovery path, which is exactly what the atomicity promise says cannot happen.

### Steps to Reproduce

1. Create a skill, then run a batch of two ops on it.
2. Make the first op succeed (a patch) and the second op raise an exception (for example an OSError from a monkeypatched handler).
3. On current main the exception escapes `_skill_manage_batch`, the `finally` deletes `snap_root`, and `_rollback()` never runs.
4. The skill now holds the first op's patch. The pre-batch snapshot is deleted.

### Expected Behavior

A raised op is a mid-batch failure like any other. The batch rolls every touched skill back to its pre-batch state, removes skills the batch created, and returns the same failure shape as the clean-failure path (`failed_index`, `completed_before_failure`, and the error message).

### Actual Behavior

The exception propagates. Earlier ops stay applied, the snapshots are destroyed, and the caller gets a generic tool error with no indication the batch was half applied.

### Affected Component

N/A (CLI only)

### Debug Report

```shell
N/A. Found by reading #97295 against current main. Reproduced with a unit test that makes one op raise (test fails on main with the raw OSError, passes with the fix).
```

### Operating System

Windows 11

### Python Version

3.11

### Hermes Version

main 586672532

`tools/skill_manager_tool.py` `_skill_manage_batch` wraps each op call in `json.loads(raw)` and checks `parsed.get("success")`, but nothing guards `skill_manage(...)` itself. The `finally` that resets the gate-bypass token also does `shutil.rmtree(snap_root, ignore_errors=True)`, so an escaping exception destroys the snapshots before `_rollback()` could use them.

### Proposed Fix (optional)

Wrap the per-op `skill_manage` call in a try/except for Exception. On a raise, run `_rollback()` and return the standard failure dict with the exception type and message in the error string. BaseException (KeyboardInterrupt, SystemExit) still propagates.

Adolanium:
Treat exceptions raised by skill batch child operations as failed operations in tools/skill_manager_batch.py. Reuse the current rollback path, including retained snapshots and recovery details when restoration fails.

Fixes #97691

Validation: 11 batch tests passed via scripts/run_tests.sh. The raised-operation regression failed on main before the fix. Real-file tests cover restored original contents and retained snapshots after a failed rollback.


Adolanium:
Follow-up found while re-reading the same executor: the rollback restore itself can destroy the skill when copytree fails, because it rmtree'd the live directory before restoring and the finally then deleted the snapshots. Fixed in #97715 (issue #97714). This PR and that one touch different hunks of the same function and merge cleanly together.

Adolanium:
I updated this on the refactored main and pushed 1efef14fd0 as one commit. I ported this into the batch module and reused main's improved rollback flow, so raised operations no longer escape and discard the recovery snapshots.

11 batch tests passed via scripts/run_tests.sh. The raised-operation regression failed on main before the fix. Real-file tests cover restored original contents and retained snapshots after a failed rollback.

PR #97697, opened 2026-08-29
fix(agent): log dropped tool results at WARNING in the pre-call sanitizer

### LINKED ISSUE #97696: [Bug]: real tool results dropped by the pre-call sanitizer are only logged at DEBUG
### Bug Description

The positional pairing walk in the pre-call sanitizer (#97203) removes tool results that do not immediately follow their declaring assistant turn. That repair is required for strict providers, but it destroys real content the model asked for, on every provider.

The only trace of the loss is a DEBUG log line. DEBUG never reaches errors.log, so while the model silently answers from incomplete data nobody can see why.

### Steps to Reproduce

1. Build a transcript where a tool result is displaced past a user turn (interrupted run, replayed history).
2. Call the API. The sanitizer drops the displaced result and the model sees nothing where the result used to be.
3. Check errors.log. Nothing. The drop is DEBUG only.

### Expected Behavior

Destroying a real tool result logs at WARNING so it lands in errors.log, matching how other real anomalies in the request path are logged.

### Actual Behavior

The drop logs at DEBUG and is invisible at default log levels.

### Affected Component

N/A (all platforms)

### Debug Report

```shell
N/A. Found by reading #97203 against current main. Reproduced with a unit test (caplog at WARNING finds no record on main, finds the drop with the fix).
```

### Operating System

Windows 11

### Python Version

3.11

### Hermes Version

main 586672532

`agent/agent_runtime_helpers.py` `sanitize_api_messages`: the `dropped_positional_orphans` branch logs at DEBUG. The `added_stubs` branch also logs at DEBUG, which is right, a stub repairs an unanswered call and destroys nothing. Only the drop branch is a content loss.

### Proposed Fix (optional)

Elevate the drop branch to WARNING. Leave the stub branch at DEBUG.

Adolanium:
Log discarded positionally orphaned tool results at WARNING in the current _pair_tool_calls_positionally helper so data loss is visible in errors.log. Pairing behavior and stub logging remain unchanged.

Fixes #97696

Validation: 59 message-sequence repair tests passed via scripts/run_tests.sh. The new warning regression failed on unchanged main before applying the severity change.


Enough1122:

Reviewed `agent/agent_runtime_helpers.py` + tests for PR #97697. Elevates the dropped-positional-orphan log from DEBUG to WARNING and adds a `caplog` regression. The change is exactly as scoped: the stub log stays at DEBUG (a stub repairs an unanswered call and destroys nothing), while a dropped real result — which the model explicitly asked for — is now visible in `errors.log`. The reasoning is sound and the diff is minimal.

Non-blocking observations:

1. **WARNING may be noisy on strict providers.** For providers where orphaned results are a routine repair (the #97203 pairing walk), every affected turn now emits a WARNING into `errors.log`. That's the intent, but if the rate turns out high, a rate-limited/aggregated counter might keep the signal without log spam. Purely a tuning question.

2. **`_ra()` logger vs `caplog` filter.** The test filters `caplog.at_level(logging.WARNING, logger="run_agent")` but the assertion itself just scans records for `levelno == WARNING`. If `_ra().logger`'s name isn't exactly `run_agent`, the level filter silently doesn't apply (records are still captured globally), so the test still passes — the logger-name assumption isn't actually pinned. Consider asserting the record's `name` to lock the intended logger.

3. **Minor:** the docstring/comment block added before the call is clear and matches house style; nothing to change.

The behavior change is one line plus a comment and a test — low-risk and addresses a real observability gap. Nothing blocking.

Adolanium:
I updated this on the refactored main and pushed 0ad1e4b164 as one commit. I moved the severity change to the refactored pairing helper and retained the behavioral regression under tests/agent.

59 message-sequence repair tests passed via scripts/run_tests.sh. The new warning regression failed on unchanged main before applying the severity change.

PR #97699, opened 2026-08-29
fix: fall back to legacy com.hermes.gateway launchd service name

ytr13285056058-wq:
## Problem

On macOS, `hermes gateway status` incorrectly reports `✗ Gateway is not running` even when the gateway is healthy and supervised by launchd.

## Root cause

`get_launchd_plist_path()` and `get_launchd_label()` only look for the canonical service name `ai.hermes.gateway`. However, older revisions of `install.sh` registered the launchd service as `com.hermes.gateway`. On machines upgraded from those revisions the plist lookup comes back empty, and the remaining PID sources also miss the live gateway: the pidfile is absent, and the process-table scan deliberately excludes ancestors of the invoking CLI — which includes the gateway itself when the CLI is spawned by the gateway. All three signal sources are empty, so status reports "not running".

## Fix

Fall back to the legacy `com.hermes.gateway` plist/label whenever the canonical `ai.hermes.gateway` plist is absent (default profile only). Since `start` / `stop` / `install` / `cron` all resolve through these two functions, the one change fixes the whole command family, not just `status`.

## Tests

Added `TestLaunchdLegacyNameFallback` (4 cases): legacy fallback, canonical-wins, install-default, profiled-profile.

```
pytest tests/hermes_cli/test_gateway_service.py::TestLaunchdLegacyNameFallback -q
# 4 passed
```

### DIFF
--- a/hermes_cli/gateway.py
+++ b/hermes_cli/gateway.py
@@ -3489,10 +3489,20 @@ def get_launchd_plist_path() -> Path:
 
     Default ``~/.hermes`` → ``ai.hermes.gateway.plist`` (backward compatible).
     Profile ``~/.hermes/profiles/coder`` → ``ai.hermes.gateway-coder.plist``.
+    Legacy pre-rename installs wrote ``com.hermes.gateway.plist`` — fall back
+    to it when the canonical name is absent so status/start/stop keep working
+    on upgraded machines.
     """
     suffix = _profile_suffix()
     name = f"ai.hermes.gateway-{suffix}" if suffix else "ai.hermes.gateway"
-    return _launchd_user_home() / "Library" / "LaunchAgents" / f"{name}.plist"
+    path = _launchd_user_home() / "Library" / "LaunchAgents" / f"{name}.plist"
+    if not suffix and not path.exists():
+        legacy = (
+            _launchd_user_home() / "Library" / "LaunchAgents" / "com.hermes.gateway.plist"
+        )
+        if legacy.exists():
+            return legacy
+    return path
 
 
 def launchd_gateway_labels_for_install() -> list[str]:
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed `hermes_cli/gateway.py` + tests for PR #97699. Adds a legacy-name fallback (`com.hermes.gateway` → `ai.hermes.gateway`) to both `get_launchd_plist_path()` and `get_launchd_label()`, so upgraded machines whose launchd service kept the pre-rename label are found by status/start/stop. The four tests cover legacy-only, canonical-wins, no-plist, and profiled cases — solid coverage of the branch matrix.

Non-blocking observations:

1. **Legacy-resolution logic is duplicated across the two functions** with slightly different shapes: `get_launchd_plist_path` checks `not path.exists() and legacy.exists()`, while `get_launchd_label` checks `legacy_plist.exists() and not canonical_plist.exists()`. Functionally equivalent, but each independently re-derives the same "which service is active" decision, so they can drift (or disagree transiently if a plist is created between the two calls). A shared helper would keep them provably in lockstep.

2. **`launchd_gateway_labels_for_install()` is unchanged.** If any operation enumerates labels via that function (e.g., a stop-all/status-across-labels path), the legacy service may still be missed there. Worth confirming whether any consumer relies on the label set from that function and, if so, whether it should also include the legacy name.

3. **`get_launchd_plist_path()` is no longer a pure path function** — it now performs filesystem existence checks. For callers that write a fresh plist on first install, this is a behavior change (though on a fresh machine neither plist exists, so it returns canonical). Intentional and documented, just flagging the semantic shift.

The fix matches the described root cause (label + plist lookup both missed the legacy name) and the tests verify it. Nothing blocking.

PR #96099, opened 2026-08-27
fix(models): persist curated OpenRouter catalog to disk so picker opens fast

yavarb:
## Problem

The Hermes model picker (desktop app / TUI / web `/api/model/options`) became very slow (~16s) to open. The symptom worsens as more providers/models are configured.

## Root Cause

`hermes_cli/models.py::fetch_openrouter_models()` returns the curated OpenRouter picker list but caches only in a module-level global (`_openrouter_catalog_cache`, per-process). The picker is served by fresh processes/threads, so the in-memory cache is cold on each open — it re-downloads the full ~686KB `https://openrouter.ai/api/v1/models` catalog on every picker open.

Measured: 7.25s for the catalog download alone; ~16.39s end-to-end for `build_models_payload(load_picker_context(), refresh=False)`.

The provider-local probes (custom endpoints) are 5ms or instant-connection-refused — NOT the cause.

There is already a per-provider model-ID disk cache (`provider_models_cache.json`, TTL 3600s), but the OpenRouter *curated catalog* itself had NO disk persistence.

## Fix

Adds a disk cache for the curated catalog, mirroring the existing `_nous_recommended_disk` pattern in the same file:

- **Path**: `hermes_constants.get_hermes_home() / "cache" / "openrouter_curated_catalog.json"`
- **TTL**: 3600s (1 hour), matching the existing per-provider cache
- **Read**: on cold open, checks disk before hitting network
- **Write**: after successful live refresh
- **Bypass**: `force_refresh=True` skips the disk cache entirely (the "Refresh Models" button)

Uses atomic write (`tempfile` + `os.replace`) to prevent torn writes from concurrent processes.

## Tests

4 new tests in `tests/hermes_cli/test_models.py` (class `TestOpenRouterCatalogDiskCache`):

1. **Write-on-refresh**: live fetch writes the disk cache
2. **Cold-process disk-read**: disk cache serves when network is off (asserted via mock)
3. **Stale-TTL refetch**: expired disk cache triggers a live refetch
4. **force_refresh bypass**: `force_refresh=True` ignores the disk cache

## Before/After

| Metric | Before | After |
|--------|--------|-------|
| Cold picker open | ~16.4s | ~0.01s (disk read) |
| Catalog download | 7.25s every open | Once per hour |
| Network calls per open | 1 full catalog GET | 0 (cached) |

PR #92253, opened 2026-08-22
fix(gateway): keep first model picker open responsive on cold pricing cache

fangliquanflq:
## What does this PR do?

Normal gateway-backed model picker opens now return from process-resident pricing caches and prewarm cold pricing endpoints on a daemon thread. This removes serial provider network timeouts from the first `/model` or `/api/model/options` response path while preserving pricing on later opens and keeping explicit refresh synchronous.

### Symptom

After pricing caches are cold, the first model picker open can wait for several provider pricing requests to complete or time out. The issue reports about 25-50 seconds on a Raspberry Pi 5, while later opens in the same process are immediate.

### Impact

Gateway, dashboard, desktop, and API users can be left waiting tens of seconds before the model picker appears. The exact delay depends on provider reachability and network latency.

### Bug Cause

**Trigger:** `hermes_cli/inventory.py` / `build_model_options_payload()` enables pricing for every normal picker payload.

**Causal chain:**
1. A user opens a gateway-backed model picker with cold process pricing caches.
2. `_apply_pricing()` walks provider rows and synchronously calls `get_pricing_for_provider()` for each pricing-capable provider.
3. Slow or unreachable provider endpoints consume their network timeouts serially before the picker payload can return.

**Why it is wrong:** Best-effort display metadata is fetched on the user-visible request path even though the picker can render without it and the same process caches can be populated asynchronously.

**Working sibling / contrast:** Once the process pricing caches are warm, the same picker path is immediate. Explicit user-triggered refresh remains the path that may wait for fresh network data.

**Ruled out:** The issue's measurements show that removing custom providers and excluding most providers does not remove the delay. The traced blocking calls originate from pricing enrichment rather than the models.dev catalog load.

### Fix

- Add cache-only pricing and Nous tier lookups that never initiate provider or Portal I/O.
- Build normal picker payloads from currently cached pricing, then prewarm cold pricing data on one guarded daemon thread.
- Keep explicit refresh behavior synchronous so a requested refresh still returns fresh data.
- Add regression coverage proving a blocked cold pricing fetch cannot delay the picker response and warm cached pricing remains available without network access.

## Related Issue

Closes #92244

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `hermes_cli/inventory.py` - read cached pricing on normal picker opens and prewarm cold pricing in the background.
- `hermes_cli/models.py` - add no-I/O cache-only pricing and Nous tier lookup modes.
[...]

PR #89792, opened 2026-08-19
fix(model): memoize _load_model_overrides on config file signature

riceharvest:
## What

Memoize `_load_model_overrides()` on the config file's `(path, mtime_ns, size)`
signature.

## Why

The `/model` picker builds ~4,000 provider rows; each row calls
`_load_model_overrides() -> load_config_readonly()`. That function is already
`(mtime, size)`-cached upstream, but a cache **hit** still acquires the config
lock, stats the file, and rebuilds the cache signature. Across one warm
model-picker build that is ~4,100 redundant config reads — **~76ms, ~47% of the
build's CPU** — re-reading the same immutable `model_overrides` section that
cannot change mid-build.

The old comment declined local memoization because an `id(cfg)`-keyed layer can
serve stale overrides when CPython reuses a freed dict's address. Keying on the
config file's on-disk signature (path + mtime_ns + size) is immune to that: any
edit bumps the signature and forces exactly one reload, so overrides edits are
still picked up on the next call.

## How it works

- Hit path: one `stat` + tuple compare (no lock, no upstream call).
- Miss / file changed / file absent: one real load, then memo.
- Unreadable file: falls through to a live load each time (no caching of a
  possibly-transient state).

## Measured

Warm model-picker build (backend-equivalent env, `.env` loaded):
- before: **160ms**
- after: **112ms**

This change removes the 76ms re-read. The remaining ~33ms is residual per-row
capability work, out of scope for this focused change.

## How to test

`python -m pytest tests/agent/test_models_dev.py` — includes a new
`test_memo_reuses_result_until_config_file_changes` that asserts:
1. a memo hit returns the same value **without** calling `load_config_readonly`
   again,
2. rewriting `config.yaml` invalidates the memo and the next call returns the
   new value,
3. the module-level memo is reset per test so profiles don't leak across tests.

## Platforms tested

Linux (Fedora, Python 3.11). Pure-stdlib file `stat` + in-memory dict — no
platform-specific behavior, no new I/O, no signal/process handling.

PR #97701, opened 2026-08-29
fix(vertex): pick up rotated service-account files — signature-keyed creds cache

kshitijk4poor:
## Summary

Pattern-D fix (stale cache/state after out-of-band changes) — the remaining core gap found by a systematic audit of every file-backed in-memory cache in `agent/`, `hermes_cli/`, `gateway/`, `tools/`.

**The bug:** the Vertex credentials cache (`agent/vertex_adapter.py:_creds_cache`) is keyed on the service-account file **path alone**. Rotate the file on disk — revoke a compromised key, re-issue the service account — and the process keeps minting tokens from the OLD `Credentials` object until restart. Operators rotate keys precisely when they most need the new identity to take effect.

## Real-world impact

**WHO/WHEN:** anyone running Vertex with a service-account file who rotates it while a gateway/agent process is alive. Before: the old identity is served for the life of the process (silently — tokens still mint fine until the old SA is disabled server-side, at which point every call 401s with no local explanation). After: the next call stats the file, sees the new signature, and rebuilds credentials from the rotated file.

## The fix

Cache key: path → `(path, mtime_ns, size)` — the established house idiom (`agent/skill_utils.py:414`, `hermes_cli/config.py:3343`, and the exact shape #89792 applies to model overrides). Superseded same-path entries are evicted on insert (bounded: one Credentials per file). ADC keeps a stable `("__adc__",)` sentinel and its existing expiry/refresh handling. Stat failure degrades to the bare-path key — byte-for-byte the old behavior.

## Verification

- New tests: rotation invalidates (rewrite + mtime bump → new key, new Credentials, old entry evicted); stat-failure fallback; ADC sentinel stability
- **Mutation check:** prod file reverted to origin/main → 3 tests fail (rotation test fails exactly at the stale-key assertion); restored → 8/8 green
- ruff green; rebased on fresh main

## Pattern-D context (the audit)

Every other file-backed cache checked already carries invalidation: `skill_utils` raw-config (sig-keyed), `config.py` (sig-keyed), `skill_bundles` (max-mtime over dir+files), `moa_loop` preset (config-stamp keyed + 300s TTL runtimes), `model_metadata`/`models_dev` (TTL + ETag). The cluster's remaining fixes live in three open PRs doing the same idiom at their sites — #89792 (model overrides memo), #92253 (non-blocking pricing prewarm), #96099 (disk-persisted catalog) — each independently tested green in isolated worktrees this pass (69/69, 91/91, 83/83) and endorsed on-thread. Unlike Patterns A/C, D has no mechanical/AST-detectable signature, so there is no CI gate to ship — the deliverable is closing the gaps and the three endorsed PRs.

### DIFF
--- a/agent/vertex_adapter.py
+++ b/agent/vertex_adapter.py
@@ -19,7 +19,7 @@ under the ``vertex:`` section; env vars take precedence over config.yaml.
 import logging
 import os
 import time
-from typing import Optional, Tuple
+from typing import Any, Optional, Tuple
 
 from agent.secret_scope import get_secret as _get_secret, is_multiplex_active
 
@@ -108,6 +108,30 @@ def _refresh_credentials(creds) -> None:
     creds.refresh(auth_req)
 
 
+def _creds_cache_key(resolved_path: Optional[str]) -> Tuple[Any, ...]:
+    """Cache key for the Credentials object backing *resolved_path*.
+
+    Keyed on the service-account file's (path, mtime_ns, size) signature —
+    the house idiom (agent/skill_utils.py, hermes_cli/config.py,
+    agent/models_dev.py) — so rotating the file on disk (new key, re-issued
+    service account) is picked up on the next call instead of serving
+    tokens minted from the old identity for the life of the process
+    (Pattern D: stale cache after out-of-band change). ADC has no file to
+    fingerprint; it keeps a plain sentinel key and its existing
+    refresh/expiry handling.
+
+    A stat failure falls back to the bare path: worst case we serve the
+    cached credentials exactly as the pre-signature code did.
+    """
+    if not resolved_path:
+        return ("__adc__",)
+    try:
+        st = os.stat(resolved_path)
+        return (resolved_path, st.st_mtime_ns, st.st_size)
+    except OSError:
+        return (resolved_path,)
+
+
 def get_vertex_credentials(credentials_path: Optional[str] = None) -> Tuple[Optional[str], Optional[str]]:
     """Return a (fresh access_token, project_id) pair or (None, None) on failure.
 
@@ -119,7 +143,7 @@ def get_vertex_credentials(credentials_path: Optional[str] = None) -> Tuple[Opti
         return None, None
 
     resolved_path = _resolve_credentials_path(credentials_path)
-    cache_key = resolved_path or "__adc__"
+    cache_key = _creds_cache_key(resolved_path)
 
     try:
         cached = _creds_cache.get(cache_key)
@@ -153,6 +177,15 @@ def get_vertex_credentials(credentials_path: Optional[str] = None) -> Tuple[Opti
                     scopes=["https://www.googleapis.com/auth/cloud-platform"]
                 )
             _creds_cache[cache_key] = (creds, project_id)
+            # A rotation leaves the old signature's entry behind; drop any
+            # other entries for the same path so the cache holds at most one
+            # Credentials per file (bounded, and stale identities don't
+            # linger for surprise reuse via an old key).
+            for k in [
+                k for k in _creds_cache
+                if k is not cache_key and k != cache_key and k[0] == cache_key[0]
+            ]:
+                _creds_cache.pop(k, None)
         else:
             creds, project_id = cached
 
--- a/agent/vertex_adapter.py
+++ b/agent/vertex_adapter.py
@@ -211,7 +211,11 @@ def get_vertex_credentials(credentials_path: Optional[str] = None) -> Tuple[Opti
 
         # If ADC failed (e.g. expired refresh token), try the SA file
         # before giving up — it may have been added after initial startup.
-        if cache_key == "__adc__":
+        # Keyed on the RESOLVED PATH being absent (i.e. this attempt was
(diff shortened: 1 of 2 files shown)


kshitijk4poor:
**Review-gate result (folded as the follow-up commit):**

The /simplify-code reviewer caught a real regression in my first commit: the ADC-failure → SA-retry fallback still compared `cache_key == "__adc__"` (string) after the keying change made every key a tuple — always-False, retry silently dead, and no existing test covered that path so 8/8 green didn't catch it.

Folded: the guard now tests the actual condition (`not resolved_path`) instead of a key-shape literal, so it can't rot again; plus a regression test driving the full ADC-fails → SA-file-appears → retry-succeeds path. Verified three ways: the test fails with the broken string comparison re-introduced, passes on main (whose string keys made the old comparison work — confirming this was a would-be regression my PR introduced and now closes), and 9/9 on the branch. Mutation check re-run on the final stack: main prod → 3 fail; restored → 9/9 green.

unsupportedpastels (commented):
## Security review: one medium-severity finding

### Medium — credential rotation can still reuse the old credentials object

**`agent/vertex_adapter.py:128-130`**

The cache fingerprint uses only:

```text
(path, mtime_ns, size)
```

Those attributes do not uniquely identify the credential contents. An atomic replacement can contain different credentials while retaining the previous file’s size and modification time. This occurs with metadata-preserving deployment/configuration tools and can also be deliberately produced.

I reproduced this against PR head `1e0ad60a8134a5fb64663e4d81a824a82f704410`:

```text
inode_changed=True
content_changed=True
cache_keys_equal=True
credentials_reloaded=False
```

In that case, Hermes continues using the cached private key and token despite the service-account file having been replaced. That undercuts the PR’s security objective: reliably picking up credential rotation without restarting.

The new rotation test does not cover this because it changes the file size and explicitly advances its mtime.

**Suggested remediation:** fingerprint the actual bytes, ideally reading the small JSON file once and constructing credentials from that same snapshot. This also avoids a stat/read TOCTOU window. A weaker but cheaper improvement would include device, inode, and change time, which catches ordinary atomic replacements but is not as definitive as a content digest. Add a regression test replacing the file with different, equal-length content while preserving `mtime_ns`.

### Other observations

- The ADC-to-service-account retry correction is sound and tested.
- Superseded cache entries are bounded by path.
- No new injection, path traversal, secret disclosure, or dependency exposure found in the diff.
- Focused tests: **9 passed**.
- Ruff: **passed**.
- Semgrep auto rules: **no findings**.
- GitHub’s supply-chain, OSV, Windows/macOS, and lint checks passed; the full Python test job and Nix check were still running when reviewed.

**Verdict:** narrow and directionally correct, but the stated credential-rotation guarantee remains incomplete due to the metadata-signature collision. No review or comment was posted to GitHub.

kshitijk4poor:
@unsupportedpastels Verified and fixed — thank you, this was a real hole in the PR's core guarantee.

**Reproduced your collision first** (equal-length content, `os.utime` restore, `os.replace`): `inode_changed=True content_changed=True cache_keys_equal=True` — exactly your result. The stat idiom I borrowed is right for the config caches it came from (guards a *parse*; a metadata-identical miss is harmless) and wrong for this one (guards an *identity*).

**Folded as `9ce700f`, your suggested remediation in full (the strong form):**
- Key is now `(path, sha256(content))` via a new `_read_sa_file()` that returns both the bytes and the key from ONE read; on a miss, credentials are built from that same snapshot via `from_service_account_info` — so the stat→read TOCTOU you noted is closed too (key and credentials can never describe different bytes).
- Read failure degrades to the bare-path key + the SDK's own file read — byte-for-byte the pre-signature behavior.
- Cost: one read+sha256 of ~2KB JSON per cache probe, noise next to the token mint the cache avoids.

**Your requested regression test is in:** equal-length replacement with `mtime_ns` preserved and atomic `os.replace` → asserts new cache key + new Credentials object. Mutation-checked: it FAILS against the stat-keyed commit (reproducing your `cache_keys_equal=True`) and against origin/main; 10/10 green on the branch.

Re-armed auto-merge (rebase).

kshitijk4poor:
**Final formal gates (full diff, all 4 commits) — passed, armed:**

- **Phase 2 / hermes-agent-dev conventions:** new tests write only under `tmp_path` (fixture sandboxes env vars + fake google-auth, no HERMES_HOME writes); green isolated, green under xdist `-n 2`; 32 wider vertex/auxiliary consumer tests green.
- **/simplify-code on the final stack:** one finding, folded as the last commit — after the content-digest rework, `_creds_cache_key()` had become a wrapper with zero production callers, kept alive only by its own tests. Consolidated into `_sa_snapshot()` (one helper owns bytes+key resolution for ADC / readable / unreadable); prod calls it, tests target it.
- **Mutation check re-run post-consolidation:** prod reverted to origin/main → 4 fail (both rotation classes + retry + digest); restored → 10/10. Tree clean, rebased on fresh main.

Review status: @unsupportedpastels's medium finding reproduced, fixed (content-digest key + same-snapshot construction), regression-tested per their spec, and replied above. No other reviews/comments outstanding.

Auto-merge (rebase) armed.

PR #63208, opened 2026-07-12
feat(plugins): add context-aware turn controllers and busy-safe commands

0error0warning:
## Summary

Adds generic plugin ABI needed by standalone long-running controllers, without any Supergoal-specific code:

- context-aware slash commands with `CommandContext.session_id` and `enqueue_followup`
- async-safe plugin command dispatch
- post-turn `TurnControlContext` / `TurnDirective` controllers with dedupe and monotonic state versions
- `on_session_rotate` hook for compression continuity
- busy-safe plugin control subcommands that never interrupt an active agent
- CLI, Gateway, TUI, and compression integration

## Design

The core remains product-name-agnostic. A standalone plugin owns state, policy, evidence, and command semantics; Hermes only provides generic lifecycle and queue primitives.

## Verification

- `tests/hermes_cli/test_plugin_turn_control_abi.py`
- `tests/gateway/test_plugin_turn_control_abi.py`
- `tests/hermes_cli/test_plugins.py`
- 133 focused Core tests passed on current `main`
- external standalone consumer suite: 57 tests passed against this branch
- `py_compile` and `git diff --check` passed

No new model tools or user-facing environment variables are added.

PR #97707, opened 2026-08-29
feat(plugins): dispatch busy-safe controls during active turns

chrsCRT:
## Summary

This is a focused current-main successor for the busy-safe command subset of #63208. It deliberately leaves the broader command-context and turn-controller ABI to a separate follow-up.

- lets plugin commands declare exact `busy_safe_subcommands` (including `""` for the bare command)
- routes opted-in controls through the real active-adapter busy callback
- preserves slash access control on both cold and busy paths
- rejects undeclared verbs mid-turn without queueing or interrupting the active run
- supports sync and awaitable handlers and marks inline replies notify-worthy
- forwards the metadata through the context-engine collector
- documents the public registration and runtime contract

The first commit preserves the original contributor as author; the gateway completion commit retains co-authorship.

## Why a focused successor

The original ABI branch is now conflicting with current `main`, and its broader session-identity/controller surface overlaps other open work. The concrete downstream blocker only requires the generic busy-safe plugin command contract. Keeping this PR to that contract avoids rebasing unrelated turn-controller APIs and makes the production-path fix independently reviewable.

## Verification

- `99 passed` across the new regression suite plus adjacent plugin, context-engine, cleanup, and slash-access suites
- Ruff: PASS
- Python compile checks: PASS
- regression sabotage: representative tests fail on unchanged upstream `main` with `unexpected keyword argument 'busy_safe_subcommands'`
- external acceptance consumer: `cortadito-ops-control` passes `hermes plugins doctor --ci`
- external active-adapter probe: `/cortadito_status` dispatches inline while active, produces one notify-worthy response, and creates no queued follow-up

- broader affected suite: `6,668 passed`, `34 skipped`, `2 xfailed`; seven order-dependent failures all passed when rerun in isolation on both this branch and the exact upstream base
- independent read-only audit bound to `415ee6974c15b193658f363510b924a0296a59c4`: PASS, no BLOCK/HIGH/MEDIUM findings

## Behavior and compatibility

- Existing plugin commands that omit `busy_safe_subcommands` keep the prior busy queue/interrupt path.
- An empty declaration opts into nothing.
- The first argument token is matched case-insensitively; `""` matches a bare command.
- Once a command opts in, undeclared verbs are rejected while a turn is active.
- Current open work in #78016 owns `pre_gateway_dispatch` parity for the generic busy path; this PR does not duplicate that unrelated fix.

## Scope exclusions

- no `CommandContext` or `TurnControlContext`
- no post-turn controllers or follow-up scheduling
- no session-identity ABI changes
- no consumer-specific logic in Hermes core

Related: #63208

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -2750,6 +2750,7 @@ from gateway.platforms.base import (
     EphemeralReply,
     MessageEvent,
     MessageType,
+    _mark_notify_metadata,
     _prefix_within_utf16_limit,
     _reply_anchor_for_event,
     build_auto_tts_output_path,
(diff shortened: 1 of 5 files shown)


alt-glitch:

Related: #63208 is the broader context-aware turn-controller ABI. This focused current-main successor supplies only the busy-safe plugin-command subset, so it is related rather than a duplicate.

chrsCRT:
Hi maintainers, this focused PR is ready for review and would unblock a downstream plugin consumer that depends on the busy-safe command ABI. When convenient, could you please review it for merge? Thank you.

PR #97158, opened 2026-08-28
fix(cache): honor the host-declared conversation key on the affinity-key path

JoaoMarcos44:
## What does this PR do?

Lets the **host declare the logical conversation** behind its physical session ids, and makes the four conversation-affinity surfaces consult that declaration. This is the fix for the key churn demonstrated in #96811 — the mechanism behind the cost half of #96570.

## Root cause

Every affinity hint Hermes sends is derived from the **physical** session id:

| Consumer | Wire field | Derived from |
| --- | --- | --- |
| `agent/transports/codex.py`, `agent/transports/chat_completions.py` | `prompt_cache_key` | `resolve_prompt_cache_scope()` → compression-lineage root → physical id |
| `plugins/model-providers/openrouter` | `body.session_id` (sticky routing) | ambient conversation id → `get_conversation_root()` → physical id |
| `plugins/model-providers/nous` | `body.session_id` (sticky routing) | same |
| `plugins/model-providers/openrouter` | `x-grok-conv-id` header | same |

A host that mints one physical session per **response** therefore re-keys all four on every reply, and the conversation never lands back on the routing bucket it just warmed. Two hosts do exactly that:

1. **Hermes Studio group chat** — `groupRuntimeSessionId()` mints `gc_run_<room>_<profile>_<name>_<uuid4hex>` per reply and destroys the session afterwards.
2. **`POST /v1/responses` with client-managed history** — `session_id = stored_session_id or str(uuid.uuid4())` in `gateway/platforms/api_server.py`. Without `previous_response_id` that is a fresh id **per request**, even though the same handler parsed `X-Hermes-Session-Key` a few lines earlier and passes it straight to `_create_agent(gateway_session_key=…)`.

Case 2 closes the whole loop inside this repo: the stable identity is already in hand, already on the agent — the affinity path just never consulted it.

## Why not normalize the id

Stripping a per-run-looking suffix is not identity-safe, and #96768 already carries both negative controls as tests: `POST /v1/sessions` preserves a client-supplied id verbatim (two independent conversations may differ only in a trailing 32-hex identity), and Studio truncates its semantic prefix to 96 characters **before** appending the token (two members of one room can end up distinguished by that token alone). Same failure class as #79017: syntactic inference truncates legitimate identities and collides distinct owners.

[...]

Issue #96811, opened 2026-08-28
Per-response session ids churn every conversation-affinity key (prompt_cache_key, sticky session_id, x-grok-conv-id)


JoaoMarcos44:
## Summary

A host that mints one physical `session_id` per **response** re-keys every conversation-affinity hint Hermes sends, so the conversation never lands back on a warm prompt prefix. Hermes has no way today to learn the logical conversation behind those ids, and it must not infer one from the id's syntax.

## Affected surfaces

Every affinity hint is derived from the physical session id through one normalizer (`agent/transports/codex.py::_cache_scope_from_session_id`):

| Consumer | Wire field | Purpose |
| --- | --- | --- |
| `plugins/model-providers/openrouter` | `body.session_id` | OpenRouter sticky routing key |
| `plugins/model-providers/nous` | `body.session_id` | Nous Portal sticky key |
| `plugins/model-providers/openrouter` | `x-grok-conv-id` header | pins xAI's prompt cache to one backend |
| `agent/transports/codex.py`, `agent/transports/chat_completions.py` | `prompt_cache_key` | OpenAI / Codex cache routing |

## Reproduction

Hermes Studio's group chat mints and destroys one session per reply:

```ts
// hermes-studio - group-chat/agent-clients.ts
export function groupRuntimeSessionId(roomId, profile, name) {
  const prefix = `gc_run_${safeId(roomId)}_${safeId(profile)}_${safeId(name)}`.slice(0, 96)
  return `${prefix}_${randomUUID().replace(/-/g, '')}`   // per-response nonce
}
```

The rows carry no lineage (`create_session(id, source, model)` - no `parent_session_id`), so `resolve_prompt_cache_scope()` correctly returns the physical id and all four keys differ for reply *N+1*.

Recorded as executable controls in `tests/agent/test_prompt_cache_scope.py::TestPerResponseRunNonceIsolation` (PR #96768): `test_affinity_keys_churn_per_response_today` and `test_provider_sticky_key_churns_per_response_today` assert the churn on current `main`; `test_lineage_walk_cannot_resolve_these_rows` shows no semantic owner resolves it.

## Why this cannot be fixed by normalizing the id

Stripping a trailing per-run-looking token (the shape tried in #96768) is not identity-safe:

1. `POST /v1/sessions` preserves a client-provided `body["id"]`/`body["session_id"]` verbatim, so two independent conversations may legitimately share a prefix and differ only in a trailing 32-hex identity.
[...]

PR #96768, opened 2026-08-28
test(cache): pin the prompt-cache scope isolation invariant for per-response session ids

JoaoMarcos44:
## What does this PR do?

Pins the **cache-scope isolation invariant** for hosts that mint one physical session per response, and records the per-response affinity-key re-keying those hosts produce.

**This PR owns two independent test contracts, deliberately:**

| Contract | File | What it pins |

| --- | --- | --- |

| Cache-scope isolation (the title) | `tests/agent/test_prompt_cache_scope.py` | distinct physical ids never merge onto one affinity key; cron's per-fire timestamp stays the only carve-out |

| System-prompt persistence under a per-response session | `tests/agent/test_system_prompt_restore.py` | the #96570 lifecycle clarification — the null-prompt warning is a first-turn artifact, not a lost write |

They share issue #96570 as their origin, not a mechanism. The evidence is the routing/affinity key changing per response; **no provider cache telemetry or billing outcome is measured or claimed anywhere in this PR.**

This PR started as a production change to `agent/transports/codex.py` (strip a trailing UUID4 hex from the logical cache scope). Review showed that rule is not identity-safe, so **that hunk is reverted** and only tests remain. The demonstrated key-churn contract now has its own issue: #96811.

### Why the id-normalization approach was dropped

Two negative controls, both from @cervantesh's review, are now executable tests:

1. **External client-supplied identities.** `POST /v1/sessions` preserves `body["id"]` / `body["session_id"]` verbatim, so two independent conversations may legitimately share a prefix and differ only in a trailing 32-hex identity. Stripping it merges them and makes their `prompt_cache_key` identical.
2. **Studio's own truncation boundary.** `groupRuntimeSessionId()` slices the semantic prefix `gc_run_<room>_<profile>_<name>` to 96 characters **before** appending the per-response token, so two members of one room whose names diverge past that boundary are distinguished *only* by that token. There is nothing left to key on once it is removed — which also shows the churn cannot be repaired from the id alone.

This is the same failure class recorded during the logical-scope design in #79017: speculative id-regex normalization truncates legitimate identities and collides distinct owners. The accepted design resolves scope from semantic lineage, and the host that knows the logical conversation has to declare it.

### What the churn issue still is

Every conversation-affinity hint Hermes sends is derived from the physical session id through one normalizer, so a per-response id re-keys all four on every reply:

| Consumer | Wire field | Purpose |
| --- | --- | --- |
[...]

Issue #96570, opened 2026-08-27
Group chat sessions: stored system prompt is null every turn — rebuilt from scratch, prefix cache always misses


Code0020161314:
Environment: Windows 11 · Hermes Studio 0.6.47 · hermes runtime 0.20.4

## Problem

Every group-chat run logs:

```
Stored system prompt for session gc_run_* is null; rebuilding from scratch this turn.
Prefix cache will miss until the rebuild persists.
```

The gc run's `update_system_prompt` write path never persists, so every turn rebuilds the system prompt from scratch. Result: group chat never benefits from prefix caching — input tokens are billed at full miss rate, making GC sessions significantly more expensive than normal sessions (related to #94222).

## Request

1. Fix the gc run system-prompt persistence (`update_system_prompt` write path).
2. Add a regression test asserting the stored prompt is non-null after the first turn of a group-chat run.

cervantesh:
I traced the current Hermes Studio group-chat lifecycle, and there is an important distinction that may change the diagnosis here.

Studio generates a new `gc_run_*_<uuid>` session ID for every group-chat response:

https://github.com/EKKOLearnAI/hermes-studio/blob/4ae23c69bbf29a287206b9b2cc9def4b5e14cdf4/packages/server/src/modules/studio/services/group-chat/agent-clients.ts#L1269-L1273

That temporary session is destroyed after the response completes:

https://github.com/EKKOLearnAI/hermes-studio/blob/4ae23c69bbf29a287206b9b2cc9def4b5e14cdf4/packages/server/src/modules/studio/services/group-chat/agent-clients.ts#L1517-L1524

Before `run_conversation()` begins, the bridge also creates the SQLite session row and pre-persists the user message. The row therefore starts with a null system prompt during the same run:

https://github.com/EKKOLearnAI/hermes-studio/blob/4ae23c69bbf29a287206b9b2cc9def4b5e14cdf4/packages/server/src/modules/hermes/services/bridge/python/bridge_pool.py#L1495-L1547

Hermes then reads that new row, emits the warning because history is already present, builds the prompt, and calls `update_system_prompt()`. At that point the row already exists, so the current `UPDATE` should match and persist the prompt:

https://github.com/NousResearch/hermes-agent/blob/0dfba37b11ff2ca908ae2df85b55f4f4c9b7fd8b/agent/conversation_loop.py#L1028-L1078

[...]

PR #97709, opened 2026-08-29
fix(cache): honor the host-declared conversation key on the affinity-key path (salvage of #97158)

kshitijk4poor:
## Summary

Salvage of PR #97158 by @JoaoMarcos44 onto current `main`, authorship preserved, plus a one-line fix for the CI failure the PR branch carried.

Lets the host declare the logical conversation behind its physical session ids, and makes all four conversation-affinity surfaces consult that declaration: `prompt_cache_key` on both OpenAI-wire transports, the OpenRouter/Nous sticky `session_id`, and xAI's `x-grok-conv-id`. Hosts that mint one physical session per RESPONSE (Hermes Studio group chat, `POST /v1/responses` with client-managed history) re-keyed all four on every reply, so the conversation never landed back on the routing bucket it warmed (#96811, the cost half of #96570). The declared key is the already-in-hand `gateway_session_key` (`X-Hermes-Session-Key` / `build_session_key`), hashed to `gwk_<sha256[:24]>` so no platform/chat/user identifier reaches a provider. Explicit fork children (`/branch`, delegate, tool) and background-review forks keep their own scope (#79161); no id-syntax inference (#79017's failure class stays closed — see #96768's negative controls).

## Changes

- Contributor commit (verbatim): `agent/portal_tags.py` affinity-scope ContextVar, `agent/prompt_cache_scope.py` `declared_conversation_scope` (memoized, fork-boundary-aware, degrades to physical id on DB error), `hermes_state.py` public `is_explicit_fork_child`, openrouter/nous profile fallback chains, `run_agent.py` scope publication for turns and out-of-turn compaction, 365-line test file
- Follow-up commit (ours): initialize `affinity_token = None` alongside `token = None` in `run_conversation` — the turn-lease timeout/interrupt early paths return before `set_affinity_scope()` runs, and the `finally` reads the name unconditionally → `UnboundLocalError`. This was exactly the 4 red cross-process lease tests on the PR's CI; reproduced locally, red→green with this init.

## Review notes

- Memoization verified: the fork-check DB read happens once per session segment, not per API call; no gwk_/physical oscillation path (error → not memoized → retried until success; lineage hit → pinned).
- Mid-conversation upgrade: existing gateway conversations take ONE provider-side affinity re-key on first response after upgrade, then stay stable — strictly better than the per-response churn being fixed.
- Known limitation (documented, not blocking): `gwk_` is an unsalted hash of the chat key, so two installs serving the same chat share a sticky bucket — pre-existing property of `build_session_key`, edge case.
- Coexists with #96768's isolation invariants (its agents declare nothing → unchanged fallback path).

## Validation

| Suite | Result |
|---|---|
| test_declared_conversation_scope + test_prompt_cache_scope + transports/test_chat_completions + test_cross_process_turn_lease + test_run_agent | 390 passed |
| PR branch CI (before fix) | 4 failed in the lease suites |
| This branch (after one-line init) | 12/12 lease tests green |

Closes #97158. Refs #96811, #96570.

### DIFF
--- a/agent/portal_tags.py
+++ b/agent/portal_tags.py
@@ -56,6 +56,48 @@ _conversation_id: ContextVar[Optional[str]] = ContextVar(
 )
 
 
+# ── Ambient routing/affinity scope ───────────────────────────────────────────
+#
+# Separate from the conversation id above, which is an ATTRIBUTION value: it
+# names the conversation a request belongs to and is sent to the Portal as
+# ``conversation=<id>``. The affinity scope is a ROUTING value — OpenRouter's
+# sticky ``session_id``, Nous Portal's sticky key and xAI's ``x-grok-conv-id``
+# use it to pin one conversation to one backend/prompt cache.
+#
+# The two agree for every host that keeps one session id per conversation, so
+# the providers historically read the attribution id for both. They diverge
+# for a host that mints one physical session per RESPONSE: attribution still
+# resolves per row, while routing must follow the key the host declared for
+# the whole chat (``agent.prompt_cache_scope.declared_conversation_scope``,
+# issue #96811). Only that declared value is published here — unset means
+# "no declaration", and consumers fall back to the conversation id exactly as
+# before, so delegate trees keep sharing their parent's sticky key.
+_affinity_scope: ContextVar[Optional[str]] = ContextVar(
+    "hermes_affinity_scope", default=None
+)
+
+
+def set_affinity_scope(scope: Optional[str]):
+    """Publish the declared routing/affinity scope for this turn.
+
+    Returns the ContextVar token; pair with :func:`reset_affinity_scope`.
+    """
+    return _affinity_scope.set(scope or None)
+
+
+def reset_affinity_scope(token) -> None:
+    """Restore the previous affinity scope (pair with ``set_affinity_scope``)."""
+    try:
+        _affinity_scope.reset(token)
+    except Exception:
+        _affinity_scope.set(None)
+
+
+def get_affinity_scope() -> Optional[str]:
+    """Return the declared routing/affinity scope, or ``None`` when unset."""
+    return _affinity_scope.get()
+
+
 def set_conversation_context(conversation_id: Optional[str]):
     """Publish the active conversation id for ambient Portal tagging.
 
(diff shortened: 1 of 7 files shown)


kshitijk4poor:
Status: converting to draft — @cervantesh's review on #97158 contains one blocker this salvage fixes and one it does not.

**Fixed here:** the cleanup regression (blocker 2). `affinity_token` is initialized alongside `token`, and the 4 previously-red cross-process lease tests are green on this branch (12/12).

**Verified still present (blocker 1):** the `/new` lifetime violation. Reproduced on this exact head with the resolver + a real `SessionDB`:

```python
db.create_session('sess-A', source='telegram')
a = agent(session_id='sess-A', _gateway_session_key=K)
db.create_session('sess-B', source='telegram')   # what /new produces — same chat key K
b = agent(session_id='sess-B', _gateway_session_key=K)
resolve_prompt_cache_scope(a)  # gwk_e545a9eb80a9283345539b21
resolve_prompt_cache_scope(b)  # gwk_e545a9eb80a9283345539b21  ← does NOT rotate
```

`gateway/session.py::reset_session()` mints a fresh physical id but keeps the chat key, so declared-key precedence maps pre- and post-`/new` conversations onto one affinity scope — violating the #79017/#86733 contract (continuity across compression rotation, reset on `/new`/`/branch`), and doing so for every native messaging gateway, not just per-response hosts.

**What a correct carrier needs** (matching the review's criteria): stable across per-response physical ids within one logical conversation, rotating on `/new`. The chat key alone cannot provide that — it needs a per-conversation epoch mixed in (e.g. a value minted at `reset_session()`/conversation creation and carried alongside the key), which is a design change to the declaration contract, not a follow-up commit.

Holding this salvage as draft until the lifecycle question is settled on #97158. The `affinity_token` fix commit here remains valid for whichever revision lands.

kshitijk4poor:
Closing this salvage draft alongside #97158: the lifecycle blocker (declared chat-key scope does not rotate on `/new` — reproduced on this exact head, see the draft note above) requires a revision of the declaration contract itself (key + per-conversation epoch, covering the `/v1/responses`/Studio boundaries), so the eventual fix will be a fresh PR implementing that contract rather than commits stacked on this branch. @JoaoMarcos44's commit and the `affinity_token` cleanup fix here will be carried into that revision with authorship preserved.

Issue #97681, opened 2026-08-29
Let Bots collaborate across gateways


dokterdok:
# Let Bots collaborate across gateways, independently of Desktop

Create a Group Chat, let Bots on different gateways work together, and pick it up from another device without keeping Desktop open. Bots retain their own models, tools and credentials.

![Bots on a laptop, home server and VPS collaborating in one Group Chat, with Desktop and phone access](https://raw.githubusercontent.com/dokterdok/hermes-agent/653a19574bd6b5de7fa0cbb1ce67a057446b6c51/docs/issue-assets/bot-group-chats-concept.png)

## For reviewers

This is an umbrella roadmap, not a request to merge every linked PR.

For a bounded starting point, **#99960** addresses a lost peer response being mistaken for definite task rejection. It targets #106742 without requiring the unpublished Files/Output stack; maintainer review and CI remain separate gates.

The broader Route and Files changes are public drafts with explicit dependencies. The tables below distinguish current review units from earlier contributions and alternatives. **#106742 does not need to wait for this whole roadmap.**

## Seven layers: what each one enables

The foundations remember work, run it on the gateway and connect gateways. Clients and feature tracks build on those contracts. The numbers explain responsibilities—not a rule that every earlier feature must finish first.

| Layer | User benefit |
| --- | --- |
| **1. Durable identity and replay** | Return to the same ordered conversation and recognize work already accepted. |
| **2. Gateway execution** | Keep the group running after Desktop closes. |
| **3. Cross-gateway collaboration** | Let Bots on different gateways request work and exchange results. |
| **4. Desktop UX** | See progress, reconnect, send follow-ups and control hosted groups. |
| **5. Files and outputs** | Share inputs and retrieve exact versions of generated files. |
| **6. Messaging controls** | Control the group from an authorized messaging chat. |
| **7. Host-loss resilience** | Preserve and reconcile accepted work, then recover safely after losing a host. |

**Files and Messaging are parallel capabilities. Closing Desktop is not losing the gateway:** ordinary hosted work still needs that host; host-loss recovery is a separate safety problem. Shared runtime prerequisites are not extra layers.

[...]

Issue #95163, opened 2026-08-26
[Feature]: Opt-in backend-hosted group rooms — gateway-side round driver + authoritative room log


angel12:
### Problem or Use Case

Bot Mode group rooms are orchestrated entirely by the desktop renderer: `runGroupChatRounds` / `runGroupChatMemberTurn` in the hermes-bots plugin drive the serial round-robin, and the authoritative room log + watermarks live in plugin `localStorage`. Three consequences:

1. **Rooms stall when the driving client goes away.** Backgrounding or closing the desktop mid-round stops turn delivery until it returns. Any future mobile client inherits this: an iOS app backgrounded mid-round parks the room.
2. **Orchestration is getting *more* socket-bound, not less.** The WS-orphan reaper now interrupts turns after socket death, which required the per-turn socket lease workaround for group member turns (#93602). A lease protects one turn; it can't keep a round advancing with no client attached.
3. **There is no authoritative multi-device room record.** The v3 `hermes-bots-groups` ui_meta projection is a great read mirror ("so mobile can show the same messages"), but it's bounded (16 messages, ~48KB) and display-oriented; full history is capped by the renderer's `trimGroupChatLog`, and the source of truth remains one device's localStorage.

This also relates to the room lag/stall class in #92760.

### Proposed Solution

An **opt-in, per-room "hosted" mode** where one gateway owns the room — log, watermarks, and the round driver — while today's client-driven mode stays the default and fully supported:

- **`groups.*` JSON-RPC surface** (a `HandlerRegistry` module like `methods_bot_relay.py`): `capabilities`, `list`, `create/rename/disband/set_members`, `send`, `log` (delta fetch via monotonic `since_seq`), `state` (running / on-turn / pending clarifies), `import`/`export` (migrate a localStorage room in/out).
- **Storage in `state.db`**: room tables keyed by the immutable `roomId` (honoring final id-tombstones), log rows with a per-room monotonic `seq`, watermark/stranded tables. Attachments as files, refs in the log.
[...]

Issue #91911, opened 2026-08-21
Architecture: make Bot Mode identity, capability, delivery, and cancellation one control plane


andrexibiza:
## Summary

Bot Mode currently has several individually correct protections, but no single control-plane object that owns a bot turn's identity, authority, effective runtime, delivery lifecycle, and cancellation scope.

The same bot is independently inferred from mutable or partial coordinates across:

- the canonical `Bot Chat` title and prompt-protocol probe;
- profile name / `HERMES_HOME`;
- Desktop connection and cached roster rows;
- gateway session/profile context;
- runtime model/provider state;
- toolset and MCP configuration;
- background completion ownership;
- local and remote teammate delivery.

That creates a repeating class of failures: configured intent can disagree with effective capability, display identity can disagree with the writer/transport owner, asynchronous work can lose its return path, and a UI stop can retire orchestration without cancelling the underlying work.

This issue establishes one canonical Bot Mode control plane and a staged migration that preserves current behavior until parity is proven.

## Evidence and current anchors

- #91802 merged structured `message_agent`, but its authority is still reconstructed from session title + Bot Mode probe at schema injection and dispatch.
- #91832 correctly distinguishes permanent configuration refusal with `EX_CONFIG`/78, but lifecycle state remains distributed across exit codes, service managers, logs, and UI inference.
- #91834 hardens Telegram adapter shutdown against cancellation-shielded `updater.stop()`; it is adapter lifecycle hardening, not a general user-request cancellation tree.
- #91862 proposes an authenticated profile roster with effective Bot protocol capability; it needs one canonical read model rather than another independent truth source.
- #91889 adds visible group Stop and preserves active member runtime IDs; it should converge on a shared cancellation scope.
- #90954 proves detached completion notifications can be destroyed when their session owner is not live; in-memory parking still loses them across process restart.
- #90329 proves Bot room membership and routing need source-qualified identity rather than profile-name reconciliation.
- #89455 proves persisted profile tool/MCP choices can disagree with runtime-effective capability.
- #88819 proves autonomous peer messaging must not carry a long-lived bearer credential across a cross-origin redirect.
- #73923 is the existing delivery-ledger direction and should be extended rather than creating a Bot-only second ledger.
- #88061 describes the user-facing need for per-task causal traces and reliable multi-agent execution.

## Architectural invariant

**Coordinates select; current proof objects authorize.**

[...]

PR #97712, opened 2026-08-29
feat(bot-mode): add durable Group Chat authority and replay

dokterdok:
## The user problem

Today a Bot Group Chat is still coordinated by the Desktop process that started
it. If that process closes, sleeps, or disconnects, there is no durable owner
that can safely continue scheduling turns or provide the complete ordered
history to another client.

Moving execution to a gateway without first defining ownership and replay
would create a worse failure: two reconnecting clients could schedule the same
Bot turn or commit conflicting histories.

> [!IMPORTANT]
> **Direction confirmed:** root `state.db` behind this narrow, inert `groups.*`
> surface is the authority and replay boundary. The added requirement is that
> participant gateways must continuously retain a lossless follower copy so a
> later layer can promote a survivor without losing history.

This PR adds the foundation required before Group Chats can keep working
without Desktop:

- one durable Group Chat identity and ordered event log;
- one server-issued authority identity and monotonic authority epoch;
- typed actor and event admission;
- idempotent appends and byte-bounded replay pages;
- permanent retired-ID reservations after disband;
- stale-writer rejection and lineage-proven authority changes.
- an atomic replication manifest and complete-replay validator that prove the
  current schema preserves every authority transition needed by a follower.

The driver remains disabled in this slice (`driver: false`). Existing Group
Chats keep their current Desktop-owned behavior.

## Why this is separate

This is the smallest reviewable decision behind #95163. The next stacked PR
adds the same-gateway runner. Cross-gateway transport, Desktop UX, files,
messaging control, and automatic failover stay out of this PR.

## Replication compatibility

The current schema is sufficient for passive replication without making a
follower look like a second owner:

- the room snapshot carries immutable identity/membership plus current
  authority, epoch, revision, and replay cursor;
- every event carries its authority epoch;
- every `authority.claimed` event records the previous gateway, new gateway,
  and next epoch, so the full A -> B -> C lineage is reconstructible;
- active room history is never truncated: it reaches a hard bound and asks the
  user to start a new Group Chat instead.

`room_replication_manifest()` reads the baseline and lineage in one SQLite
transaction. `validate_replica_events()` accepts only a complete contiguous log
whose claims match that lineage. The new round-trip test pages a room through
two authority transfers, serializes the manifest, validates every event, and
rejects a tampered claim.

This PR does **not** send replicas over the network or authorize takeover. The
continuous-delivery layer and the split-brain-safe promotion policy remain
separate production increments tracked in #97681.

## Contract

- Stores Group Chat identity, membership, authority, revision, and events in
  root `state.db`.
- Adds `groups.capabilities`, `groups.list`, `groups.create`, `groups.state`,
  `groups.send`, `groups.log`, and `groups.disband`.
- Derives authority from the gateway's stable install identity; clients cannot
  choose or impersonate the owner.
- Accepts client ingress only as server-authored user messages. Clients cannot
  forge gateway or member events.
- Fences sends and disband against the local gateway identity and captured
  authority epoch, rejecting stale writers without mutating the log or room.
- Maps client retry keys into a server-owned user-event namespace, so a client
  cannot occupy deterministic control-event IDs.
- Bounds active Group Chats, per-Group Chat event count and bytes, gateway-wide
  event storage, retained disband payloads, list pages, and the complete
  serialized replay page while keeping a compact permanent retired-ID registry.
- Serializes install-identity creation and corrupt-file repair across processes
  on POSIX and Windows.
- Runs disk-backed RPC work away from the WebSocket reader thread without
  adding Group Chat code to `tui_gateway/server.py`.

## Compatibility

- Existing Desktop-driven Group Chats are unchanged.
- Older gateways return method-not-found and remain on the current path.
- No Bot turn starts in this PR.
- No Desktop projection, cross-gateway transport, file, or messaging behavior
  changes in this PR.

## Validation

Current exact head `ad949774f0`, two commits directly on `main@4f22543509`:

- `scripts/run_tests.sh` over the hosted-room, RPC, install-identity, and
  dashboard compatibility suites: **229 passed, 1 skipped**.
- Focused authority/event-ID/replay/replication suite after independent
  adversarial review: **56 passed**.
- Ruff on all changed Python files: passed.
- `git diff --check`: passed.
- Exact-head CI, Docker, and Nix: green on the current commit.

The current tests cover foreign-owned send/disband with no mutation, the exact
`system:room-disbanded` client-ID collision, two-/three-/four-byte Unicode,
complete replay-page encoding overhead, replication lineage, tampered-claim
rejection, monotonic cursor progress, idempotency,
age/count/byte retention pressure, reopen, stale traffic, migration, and
Windows first-open races.

The rewritten history removes every known-bad intermediate commit.
`tui_gateway/server.py` is unchanged; the install-identity extraction reduces
`hermes_cli/web_server.py` by 53 lines.

No screenshot is included because this foundation adds no user-visible
surface.

## Related work

- #97681: user outcome, review stack, and field-testing plan
- #95163: gateway-hosted Group Chat design record
- #94726 section 5: Bot Mode Group Chat lifecycle and stall class
- #89995 and merged #91279: Group Chat visibility and bounded Desktop projection
- #91911: identity, delivery, and cancellation direction

## Type of change

- [ ] Bug fix
- [x] New feature foundation
- [ ] Security fix
- [ ] Documentation update
- [x] Tests
- [x] Refactor

### DIFF
--- a/gateway/hosted_rooms.py
+++ b/gateway/hosted_rooms.py
@@ -1,0 +1,1456 @@
+"""Durable state for gateway-hosted Bot Mode rooms.
+
+This module owns only hosted-room identity and its append-only event log. It
+does not deliver events, lease relay work, or run agent turns; those concerns
+belong to the relay and the future hosted-room driver. Keeping that boundary
+explicit lets the room log compose with a durable relay without creating a
+second transport queue.
+
+The caller supplies the database path so tests and alternate gateway layouts
+can isolate state. Production handlers use the gateway's root ``state.db``.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import re
+import sqlite3
+import time
+from contextlib import contextmanager
+from pathlib import Path
+from typing import Any, Iterator
+
+
+PROTOCOL_VERSION = 2
+MAX_ROOM_ID_CHARS = 128
+MAX_EVENT_ID_CHARS = 128
+MAX_ROOM_NAME_CHARS = 200
+MAX_EVENT_KIND_CHARS = 64
+MAX_ACTOR_ID_CHARS = 128
+MAX_ACTOR_LABEL_CHARS = 200
+MAX_MEMBERS = 128
+MAX_MEMBERS_JSON_BYTES = 128 * 1024
+MAX_EVENT_JSON_BYTES = 256 * 1024
+MAX_LOG_LIMIT = 500
+MAX_LOG_PAGE_BYTES = 2 * 1024 * 1024
+MAX_ROOM_LIST_LIMIT = 500
+MAX_ACTIVE_ROOMS = 256
+MAX_DISBANDED_ROOM_TOMBSTONES = 512
+DISBANDED_ROOM_RETENTION_SECONDS = 90 * 24 * 60 * 60
+MAX_EVENTS_PER_ROOM = 50_000
+MAX_ROOM_EVENT_BYTES = 256 * 1024 * 1024
+# Leave substantial headroom below the pre-update state.db snapshot ceiling.
+# Event accounting does not include SQLite indexes or repeated room ids, so the
+# logical budget must stay well below the physical-file limit.
+MAX_GATEWAY_EVENT_BYTES = 16 * 1024 * 1024
+CONTROL_EVENT_COUNT_RESERVE = 64
+CONTROL_EVENT_BYTE_RESERVE = 1024 * 1024
+_JOURNAL_MODE_LOCK_RETRIES = 8
+
+_IDENTIFIER_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._:-]*$")
+_EVENT_KIND_RE = re.compile(r"^[a-z][a-z0-9_.-]*$")
+_ROOM_SCHEMA_COLUMNS = frozenset({
+    "room_id",
+    "name",
+    "members_json",
+    "authority_gateway_id",
+    "authority_epoch",
+    "next_seq",
+    "event_bytes",
+    "revision",
+    "created_at",
+    "updated_at",
+    "disbanded_at",
+})
+_EVENT_SCHEMA_COLUMNS = frozenset({
+    "room_id",
+    "seq",
+    "event_id",
+    "kind",
+    "actor_json",
+    "authority_epoch",
+    "payload_json",
+    "created_at",
+})
+_RETIRED_ROOM_SCHEMA_COLUMNS = frozenset({"room_id", "retired_at"})
+
+_EVENT_KINDS_BY_ACTOR = {
+    "user": frozenset({"message.user"}),
+    "member": frozenset({"message.member"}),
+    "gateway": frozenset({
+        "member.unavailable",
+        "room.activity",
+        "turn.deferred",
+        "turn.reassigned",
+        "turn.cancelled",
+        "turn.failed",
+        "turn.settled",
+        "turn.started",
+    }),
+    "system": frozenset({
+        "authority.claimed",
+        "authority.lost",
+        "room.created",
+        "room.disbanded",
+        "room.members_changed",
+        "room.renamed",
+    }),
+}
+_ACTOR_FIELDS = frozenset({"kind", "id", "display_name", "profile", "connection_id"})
+
+
+class HostedRoomError(ValueError):
+    """Base class for invalid or conflicting hosted-room operations."""
+
+
+class RoomNotFoundError(HostedRoomError):
+    """Raised when a room does not exist or has been disbanded."""
+
+
+class RoomHistoryExpiredError(RoomNotFoundError):
+    """Raised when a retired room remains reserved after history compaction."""
+
+    reason = "room_history_expired"
+
+
+class RoomConflictError(HostedRoomError):
+    """Raised when an idempotency key is reused for different room state."""
+
+
+class EventConflictError(HostedRoomError):
+    """Raised when an event id is reused with different immutable content."""
+
+
+class AuthorityConflictError(HostedRoomError):
+    """Raised when a stale room authority attempts to mutate hosted state."""
+
+    reason = "authority_conflict"
+
+
+class AuthoritySupersededError(AuthorityConflictError):
+    """Raised when a successful authority claim was later superseded."""
+
+
+def default_db_path() -> Path:
+    """Return the gateway-wide state database for the active install."""
+    from hermes_constants import get_hermes_home
+
+    home = get_hermes_home()
+    root = home.parent.parent if home.parent.name == "profiles" else home
+    return root / "state.db"
+
+
+def local_authority_gateway_id() -> str:
+    """Return the stable server-owned identity for hosted-room authority."""
+    from hermes_cli.install_identity import get_install_id
+
+    install_id = get_install_id()
+    if not install_id:
+        raise HostedRoomError("stable gateway install identity is unavailable")
+    return _validate_identifier(
+        f"install:{install_id}",
+        label="authority_gateway_id",
+        max_chars=MAX_ACTOR_ID_CHARS,
+    )
+
+
+def _canonical_json(value: Any, *, label: str, max_bytes: int) -> str:
+    try:
+        encoded = json.dumps(
+            value,
+            ensure_ascii=False,
+            sort_keys=True,
+            separators=(",", ":"),
+        )
+    except (TypeError, ValueError, RecursionError) as exc:
+        raise HostedRoomError(f"{label} must be JSON-serializable") from exc
+    if len(encoded.encode("utf-8")) > max_bytes:
+        raise HostedRoomError(f"{label} is too large")
+    return encoded
+
+
+def _validate_identifier(value: Any, *, label: str, max_chars: int) -> str:
+    if not isinstance(value, str):
+        raise HostedRoomError(f"{label} must be a string")
+    value = value.strip()
+    if not value or len(value) > max_chars or not _IDENTIFIER_RE.fullmatch(value):
+        raise HostedRoomError(f"invalid {label}")
+    return value
+
+
+def user_event_id(client_event_id: Any) -> str:
+    """Map a client retry key into the server-owned user-event namespace."""
+    normalized = _validate_identifier(
+        client_event_id,
+        label="event_id",
+        max_chars=MAX_EVENT_ID_CHARS,
+    )
+    digest = hashlib.sha256(normalized.encode("utf-8")).hexdigest()
+    return f"user:{digest}"
+
+
+def _validate_room_name(value: Any) -> str:
+    if not isinstance(value, str):
+        raise HostedRoomError("name must be a string")
+    value = value.strip()
+    if not value or len(value) > MAX_ROOM_NAME_CHARS:
+        raise HostedRoomError("invalid room name")
+    return value
+
+
+def _validate_members(value: Any) -> tuple[list[dict[str, Any]], str]:
+    if not isinstance(value, list):
+        raise HostedRoomError("members must be a list")
+    if len(value) > MAX_MEMBERS:
+        raise HostedRoomError("too many room members")
+    members: list[dict[str, Any]] = []
+    for member in value:
+        if not isinstance(member, dict):
+            raise HostedRoomError("each room member must be an object")
+        members.append(dict(member))
+    encoded = _canonical_json(
+        members,
+        label="members",
+        max_bytes=MAX_MEMBERS_JSON_BYTES,
+    )
+    return members, encoded
+
+
+def _validate_event_kind(value: Any) -> str:
+    if not isinstance(value, str):
+        raise HostedRoomError("kind must be a string")
+    value = value.strip()
+    if (
+        not value
+        or len(value) > MAX_EVENT_KIND_CHARS
+        or not _EVENT_KIND_RE.fullmatch(value)
+    ):
+        raise HostedRoomError("invalid event kind")
+    return value
+
+
+def _optional_actor_field(actor: dict[str, Any], field: str, max_chars: int) -> str:
+    value = actor.get(field)
+    if value is None:
+        return ""
+    if not isinstance(value, str):
+        raise HostedRoomError(f"actor.{field} must be a string")
+    value = value.strip()
+    if len(value) > max_chars:
+        raise HostedRoomError(f"actor.{field} is too long")
+    return value
+
+
+def _validate_actor(value: Any, *, kind: str) -> tuple[dict[str, str], str]:
+    if not isinstance(value, dict):
+        raise HostedRoomError("actor must be an object")
+    unknown = set(value) - _ACTOR_FIELDS
+    if unknown:
+        raise HostedRoomError(f"unknown actor fields: {', '.join(sorted(unknown))}")
+
+    actor_kind = value.get("kind")
+    if not isinstance(actor_kind, str) or actor_kind not in _EVENT_KINDS_BY_ACTOR:
+        raise HostedRoomError("invalid actor.kind")
+    if kind not in _EVENT_KINDS_BY_ACTOR[actor_kind]:
+        raise HostedRoomError(f"actor kind '{actor_kind}' cannot append '{kind}'")
+
+    actor_id = _validate_identifier(
+        value.get("id"),
+        label="actor.id",
+        max_chars=MAX_ACTOR_ID_CHARS,
+    )
+    actor = {"kind": actor_kind, "id": actor_id}
+    for field, max_chars in (
+        ("display_name", MAX_ACTOR_LABEL_CHARS),
+        ("profile", MAX_ACTOR_ID_CHARS),
+        ("connection_id", MAX_ACTOR_ID_CHARS),
+    ):
+        field_value = _optional_actor_field(value, field, max_chars)
+        if field_value:
+            actor[field] = field_value
+    encoded = _canonical_json(
+        actor,
+        label="actor",
+        max_bytes=4 * 1024,
+    )
+    return actor, encoded
+
+
+def _initialize_schema(conn: sqlite3.Connection) -> None:
+    conn.execute(
+        """CREATE TABLE IF NOT EXISTS hosted_rooms (
+            room_id TEXT PRIMARY KEY,
+            name TEXT NOT NULL,
+            members_json TEXT NOT NULL,
+            authority_gateway_id TEXT NOT NULL,
+            authority_epoch INTEGER NOT NULL DEFAULT 1 CHECK (authority_epoch >= 1),
+            next_seq INTEGER NOT NULL DEFAULT 1 CHECK (next_seq >= 1),
+            event_bytes INTEGER NOT NULL DEFAULT 0 CHECK (event_bytes >= 0),
+            revision INTEGER NOT NULL DEFAULT 1 CHECK (revision >= 1),
+            created_at REAL NOT NULL,
+            updated_at REAL NOT NULL,
+            disbanded_at REAL
+        )"""
+    )
+    conn.execute(
+        """CREATE TABLE IF NOT EXISTS hosted_room_events (
+            room_id TEXT NOT NULL,
+            seq INTEGER NOT NULL CHECK (seq >= 1),
+            event_id TEXT NOT NULL,
+            kind TEXT NOT NULL,
+            actor_json TEXT NOT NULL,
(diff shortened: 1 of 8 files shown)


andrexibiza (commented):
Reviewed exact head `eea34e63d0c4b1505cb32b4e287fb9b3f384f5bd` against its recorded base `10e93c6ab958c7ec61cfc4416f4d4459e72ca8a7` and live `main@aff5125f8edf5095aef5d3d79bbbb101c95b9413` (14 commits past that base). I read the eight-file diff, the full hosted-room store and `groups.*` surface, the focused room/RPC/install-identity tests, the predecessor/supersession history, the stacked runner, the live issue contracts, and the exact-head workflow state. There were no comments or reviews on #97712 when I started.

A lot of the hard authority work from closed-unmerged predecessor #95314 survived the recut correctly: owner identity is server-derived rather than client-asserted; first-install identity publication/repair is cross-process serialized; legacy adoption advances the epoch and emits a real `authority.claimed` lineage receipt; schema upgrades are transactional; event sequence allocation and authority claims are fenced; client ingress cannot forge member/gateway events; and `driver: false` remains an honest boundary. The new store also stays below the godfile ceiling at 1,274 lines rather than putting the room implementation into `tui_gateway/server.py`. That is substantial, careful work. 🚀

I do have one P1 blocker introduced by the new `fix(bot-mode): bound hosted Group Chat history` commit: **the retention policy deletes the identity tombstone, so a disbanded immutable `room_id` can become a different room later (ABA).**

`#95163` is explicit that the authoritative tables are keyed by the **immutable `roomId` “honoring final id-tombstones.”** This PR encodes the same contract in `disband_room()` (“Permanently tombstone a room id”), and `tests/gateway/test_hosted_rooms.py::test_disband_is_idempotent_and_room_id_cannot_be_reused` correctly proves reuse is rejected while the tombstone exists.

But the new `_prune_disbanded_rooms_locked()` in `gateway/hosted_rooms.py` selects disbanded rooms after the 90-day window, after the retained-tombstone count exceeds `MAX_DISBANDED_ROOM_TOMBSTONES`, or under gateway byte pressure, then executes both:

```py
DELETE FROM hosted_room_events WHERE room_id IN (...)
DELETE FROM hosted_rooms WHERE room_id IN (...)
```

After that second delete, `create_room()` finds no existing row and takes the fresh-room INSERT path for the same `room_id`. The new `test_room_listing_is_paged_and_old_tombstones_are_pruned` actually proves the first half of this sequence by making `room-0` disappear once the tombstone bound is crossed; it never exercises the other side of the shape by attempting to create `room-0` again.

That turns bounded history retention into identity reuse. A stale client, replay cursor, future task/cancellation record, or cross-gateway envelope carrying the old room coordinate can now alias a different logical Group Chat after pruning. This is exactly the failure the authority layer is otherwise designed to prevent: a coordinate that once named terminal generation A becomes valid again for generation B with no proof object distinguishing them. It also breaks the PR-body claim that writes after disband are rejected as a durable property; they are rejected only until retention pressure erases the evidence.

**Required repair:** separate payload/history retention from identity retirement. It is fine to bound or purge old event payloads, but keep a compact permanent retired-ID record (in `hosted_rooms` or a dedicated retired-room table) that `create_room()` consults forever, so a final room identity never returns to the fresh namespace. If disband replay itself is retention-bounded, make the expired-history state typed/explicit; do not turn “history no longer retained” into “this identity never existed.” An alternative is a server-issued globally unique room identity plus a durable non-reuse guarantee, but client-selected IDs cannot be made reusable under the present contract.

Please add a regression that forces each pruning trigger (retention age, tombstone-count pressure, and byte pressure), then proves: (1) the retired ID still cannot be created again after reopen, (2) stale send/log traffic cannot alias a new logical room, and (3) an actually new room ID remains creatable. The existing positive pruning test can stay, but it should assert history compaction rather than identity erasure.

**Interlocks / landing order:** #97744 is explicitly stacked on this exact foundation and gives durable tasks, generations, cancellation, and restart recovery meaning *inside a room identity*, so it should not inherit a reusable room key. #95163 remains the direct design owner and supplies the final-tombstone requirement; #97681 supplies the product-level stale-coordinator/late-result rule; #91911 is complementary control-plane work whose “coordinates select; current proof objects authorize” rule is violated if a retired coordinate can silently select a new object. #92931 is complementary fallback relay/settlement work, not a duplicate room store, but it shares `tui_gateway/server.py`; keep one room authority and one delivery authority rather than merging historical carriers by text. Open #93508 also currently touches both `hermes_cli/web_server.py` and `tui_gateway/server.py`, so whichever of those live objects lands second needs a semantic rebase on those integration seams. Open #78630 remains the structural owner for sharding `tui_gateway/server.py`; this PR is directionally good in keeping the actual new behavior in `methods_groups.py` / `hosted_rooms.py` rather than growing another inline implementation there.

Exact-head hosted workflows are green on the reviewed SHA: [CI 33238522582](https://github.com/NousResearch/hermes-agent/actions/runs/33238522582), [Docker 33238522294](https://github.com/NousResearch/hermes-agent/actions/runs/33238522294), and [Nix 33238522325](https://github.com/NousResearch/hermes-agent/actions/runs/33238522325). Those runs certify the submitted object but do not cover the post-prune ABA sequence above. Since live main has advanced 14 commits beyond the recorded base, the repaired final head should also be recomposed on current main and receive fresh exact-head CI before #97744 treats it as a landing dependency.

Once the tombstone is made identity-preserving rather than identity-erasing, I do not see another blocker in the inert authority/replay layer represented here. The transactional authority, replay, migration, and single-winner install-identity work is strong and worth preserving.

dokterdok:
Addressed the ABA blocker from @andrexibiza's review in exact head `50ec1e1ffb068236b4bd4cb79f9d3c883d53c8b9`.

- Disband now writes a compact permanent retired-ID reservation in the same transaction.
- Age, tombstone-count, and byte-pressure pruning delete replay/worker payloads only after preserving that reservation.
- `groups.send`, `groups.log`, state/authority access, and the later rename path distinguish retained tombstones from genuinely compacted history with stable `reason: room_history_expired` data.
- Repeated disband stays idempotent after compaction; migration backfills already-disbanded rows.
- The review's reopen, stale traffic, fresh-ID, and migration cases are covered, along with deterministic native-Windows publication/SQLite first-open races found during the focused gate.

Focused Layer 1 result: **48 passed**, Ruff passed, and `git diff --check` passed. I recomposed the stack in order onto current `main@aff5125f8e`: #97712 `50ec1e1ffb` → #97744 `d1c90e9d18` → draft #97797 `0a4f724426`. The PR is temporarily draft while fresh exact-head CI reruns.

tui_gateway/methods_groups.py
```diff
@@ -0,0 +1,239 @@
...
+        default_db_path,
+        room_state,
+    )
+
+    try:
+        room = room_state(default_db_path(), room_id=params.get("room_id"))
+        event = append_event(
+            default_db_path(),
+            room_id=params.get("room_id"),
+            event_id=params.get("event_id"),
```
andrexibiza: **[P1] Domain-separate client idempotency keys from control-event IDs.**

This accepts any client `event_id`, including `system:room-disbanded`. The same room then cannot be disbanded: `disband_room()` inserts its fixed control receipt into the same `UNIQUE(room_id, event_id)` namespace, gets an integrity failure, and rolls the transaction back. Because the log is append-only, retry cannot repair the room. Server-mint/prefix the authoritative user-event ID (storing the client key separately) or reject a closed internal namespace, then add a send-with-`system:room-disbanded` → disband regression.

tui_gateway/methods_groups.py
```diff
@@ -0,0 +1,239 @@
...
+            kind="message.user",
+            actor={"kind": "user", "id": "desktop"},
+            payload=params.get("payload"),
+            authority_gateway_id=str(room["authority_gateway_id"]),
```
andrexibiza: **[P1] Do not reconstruct gateway authority from the target row.**

After `claim_authority(A → B, 1 → 2)`, stale gateway A calls `groups.send`, reads B / 2 here, and submits B / 2; `append_event()` accepts because the row matches itself. `groups.disband` repeats the same defect at lines 206–207. Derive `local_authority_gateway_id()`, reject when the stored owner is not local, then pass the local ID with the captured epoch so the store remains the race fence rather than the source of impersonated authority.

gateway/hosted_rooms.py
```diff
@@ -0,0 +1,1414 @@
...
+            _raise_room_not_found(conn, room_id)
+        latest_seq = int(room["next_seq"]) - 1
+        if since_seq > latest_seq:
+            raise HostedRoomError("since_seq is ahead of the hosted room log")
+        rows = conn.execute(
+            """WITH candidates AS (
+                   SELECT room_id, seq, event_id, kind, actor_json,
+                          authority_epoch, payload_json, created_at,
+                          SUM(
+                              LENGTH(event_id) + LENGTH(kind) +
```
andrexibiza: **[P2] Measure this window in UTF-8 bytes, not TEXT characters.**

`_canonical_json()` uses `ensure_ascii=False`, but SQLite `LENGTH(TEXT)` counts Unicode code points. Emoji/CJK payloads therefore consume 2–4 bytes while counting as one here, so the page can exceed `MAX_LOG_PAGE_BYTES` by roughly 4×. Use `LENGTH(CAST(... AS BLOB))` for all four fields (as the migration backfill already does) and add a multibyte regression that measures the serialized UTF-8 page.

andrexibiza (commented):
Re-reviewed exact head `5aead6f8750df3993c54ef3fbc20b47abefa26e2` against synthetic merge `775c263d46f5e2c2321da640e8db5e81b1854194` (base `66666f6e2eca0ae883195a34c66131985ea7dd06`) and live `main@4f22543509d1b91dc45bcb369447126c5eb14fb7`. The seven later main commits do not touch this slice. The prior post-prune room-ID ABA blocker is fixed by the permanent retired-ID registry. Exact-head CI, Docker, and Nix are green.

This head is still not landable. I have three current-object blockers and one commit-train acceptance gate.

### 1. P1 — the public mutation boundary reconstructs foreign authority from ambient state

`groups.send` reads `room_state()` and passes the row's own `authority_gateway_id` / `authority_epoch` back as the authorization proof. `groups.disband` does the same. That comparison detects a race, but it does not authenticate the gateway performing the mutation.

The deterministic sequence is:

1. create under gateway A at epoch 1;
2. `claim_authority(A → B)` advances the row to B / epoch 2;
3. invoke public `groups.send` or `groups.disband` on stale gateway A;
4. A reads B / 2 from the row and submits B / 2, so the store accepts A as B.

That directly contradicts the PR contract that authority is server-derived and clients/stale writers cannot impersonate the owner. Derive `local_authority_gateway_id()` at the RPC/service boundary, fail closed when the stored owner differs, and pass the local identity plus the captured epoch into the fenced store operation. Add RPC regressions proving foreign-owned send and disband return the typed authority conflict and leave both the event log and tombstone state unchanged. #97744 already introduces this local-owner check in its service layer; the parent foundation must be correct independently rather than relying on a child PR to repair its public surface.

### 2. P1 — client idempotency keys share the internal control-event namespace

`groups.send` forwards the client-controlled `event_id` directly into the authoritative per-room `UNIQUE(room_id, event_id)` namespace. A client can send a normal user message with `event_id="system:room-disbanded"`. `disband_room()` later tries to insert its fixed control receipt with that same ID, hits the unique constraint, rolls back, and can never disband that append-only room. The same class lets ingress pre-occupy any later deterministic internal event coordinate.

Domain-separate the client idempotency key from the authoritative log event ID: server-mint/prefix the stored user-event ID and retain the client key in its own domain, or reject a closed reserved namespace that covers every internal producer. Add the exact send-then-disband adversarial regression and prove reopen/retry cannot leave the room permanently active.

### 3. P2 — the replay "byte" window counts Unicode code points

`read_events()` uses `LENGTH(TEXT)` for `actor_json` and `payload_json`, while `_canonical_json(..., ensure_ascii=False)` stores raw Unicode and admission limits use `len(...encode("utf-8"))`. SQLite therefore counts an emoji as one unit here although it occupies four UTF-8 bytes. A nominal 2 MiB page can exceed the advertised byte bound by roughly 4×; the existing test uses ASCII only.

Use `LENGTH(CAST(... AS BLOB))` for every field in the cumulative window, matching the migration backfill already present in this file, and add a multibyte payload regression that measures the serialized UTF-8 page.

### 4. Commit-train gate — the submitted history still contains a known-bad object and lacks every-commit proof

The five-commit train retains `47754e592ec9f534fe75ede381912dcfe03c0a57`, where bounded pruning introduced the room-ID ABA defect fixed only by `50ec1e1ffb068236b4bd4cb79f9d3c883d53c8b9`. The first two commits (`a3a6e5b303914219748ac09c2d9a90fcdab967fe`, `47754e592ec9f534fe75ede381912dcfe03c0a57`) also have no exact hosted workflow runs; only the later three commits have CI/Docker/Nix proof. A green final head does not make a known-bad intermediate Git object green.

Fix up/squash the repairs into a coherent history, rebase the repaired parent on current main, and obtain fresh CI/Docker/Nix proof for every remaining commit. Because #97744 and #97797 are stacked on this exact lineage, they must then be restacked on the repaired parent and reacquire their own every-commit proof.

Current-head receipts: [CI 33315638726](https://github.com/NousResearch/hermes-agent/actions/runs/33315638726), [Docker 33315638141](https://github.com/NousResearch/hermes-agent/actions/runs/33315638141), [Nix 33315638134](https://github.com/NousResearch/hermes-agent/actions/runs/33315638134). Those runs certify the submitted merge object; they do not exercise the three adversarial paths above.

andrexibiza (commented):
Re-review of exact head `5aead6f8750df3993c54ef3fbc20b47abefa26e2`.

The prior retired-room ABA blocker is repaired on this head: history compaction now preserves a permanent retired-ID reservation, and the stale-history state is typed. The exact-head CI, Docker, and Nix workflows are green.

### P1 — `MAX_LOG_PAGE_BYTES` is enforced in characters, not UTF-8 bytes

`_canonical_json()` stores JSON with `ensure_ascii=False` and validates its real UTF-8 byte length. `read_events()`, however, computes `cumulative_bytes` with SQLite `LENGTH(event_id) + LENGTH(kind) + LENGTH(actor_json) + LENGTH(payload_json)`. For TEXT values SQLite `LENGTH()` counts Unicode characters. A page containing multibyte payloads can therefore cross the advertised 2 MiB replay ceiling by several times while still satisfying the SQL predicate.

This is a boundary violation, not only an accounting mismatch: peer/Desktop replay callers rely on the page byte cap to bound allocation and transport. The existing tests exercise ASCII and do not prove the non-ASCII side of the contract.

Required repair:

- account bytes in SQLite, e.g. `LENGTH(CAST(payload_json AS BLOB))` and the same treatment for every text field in the sum, or persist exact encoded-byte counts;
- add a multibyte regression using two-/three-/four-byte code points that asserts the serialized page remains within `MAX_LOG_PAGE_BYTES`, preserves monotonic cursor progress, and eventually returns all events.

### Hard source invariant

This PR modifies `tui_gateway/server.py` and `hermes_cli/web_server.py`, both already above the repository's 2,000-physical-line ceiling. A touched file must remain below that bound; the registration/adaptation seams need to be extracted rather than extended in-place.

This head is not merge-ready until the byte-bound defect and touched-godfile violations are repaired and the replacement exact head receives fresh proof.

andrexibiza (commented):
Re-reviewed exact head `f5908c18d86a4d4a572a464be2d00e9eb8cec3ee`, which is now one commit directly parented on `main@4f22543509d1b91dc45bcb369447126c5eb14fb7`. The rewritten object does materially repair the prior behavioral blockers: `groups.send`/`groups.disband` derive the local gateway identity and reject a foreign stored owner before mutation; client retry keys are domain-separated through `user_event_id()`; replay accounting uses `LENGTH(CAST(... AS BLOB))`; the permanent retired-ID registry preserves room identity after history compaction; and the exact-head CI/Docker/Nix runs are all green (`33332784224`, `33332783891`, `33332783919`). The known-bad intermediate history is also gone.

One acceptance blocker remains under the repository 2K invariant: this exact commit still modifies `hermes_cli/web_server.py` (`+4/-57`). The extraction is directionally correct and shrinks the file, but a touched file that remains above the 2,000-physical-line ceiling does not satisfy the gate. The PR already introduces `hermes_cli/install_identity.py`; finish the seam so this slice no longer submits a modified >2K godfile (or complete the owning extraction needed to bring the touched file below the ceiling), then reacquire exact-head proof on the replacement object.

I do not see another blocker in the hosted-room authority/replay mechanics on this head. The remaining issue is structural acceptance, not the prior authority, namespace, replay-byte, or tombstone behavior.

andrexibiza (commented):
Live-main correction to my immediately preceding review: `main` advanced during the review window to `d63f996a757f6255fc1454239616ab4b4435e0f5`, two commits beyond this head's parent `4f22543509d1b91dc45bcb369447126c5eb14fb7`. I compared that exact interval. The new main paths are Photon/docs-only (`plugins/platforms/photon/**`, Photon tests, and Photon/environment docs) and are path-disjoint from this PR; GitHub currently reports #97712 mergeable.

That does not reopen any of the hosted-room behavioral findings I marked repaired. It does mean the final acceptance statement must be narrower: `f5908c18d8` has a green exact-head matrix on its submitted base, not a current-main exact-object proof. The replacement that closes the remaining 2K touched-file gate should also be recomposed onto landing main and receive fresh CI/Docker/Nix. I am recording this separately so the public review history does not leave the earlier "current main" wording stale.

andrexibiza (commented):
Re-reviewed moved head `ad949774f0072499c56c8e0e86fb056cb877b315` against parent `f5908c18d86a4d4a572a464be2d00e9eb8cec3ee`, submitted base `4f22543509d1b91dc45bcb369447126c5eb14fb7`, and live `main@d63f996a757f6255fc1454239616ab4b4435e0f5`.

The new commit is narrowly additive: `+163` in `gateway/hosted_rooms.py` and `+74` in `tests/gateway/test_hosted_rooms.py`. It adds the room replication manifest/lineage validator and a paged A→B→C authority-transfer round trip. I rechecked the earlier authority, client-event namespace, UTF-8 replay-byte, retired-ID ABA, and stale-writer repairs on the full current diff; those fixes remain intact. `gateway/hosted_rooms.py` is 1,619 lines and the new test file is 1,092 lines, so the new replication layer itself stays below the 2K ceiling.

The surviving two-commit train now has exact hosted proof on every object. `f5908c18...` is green in CI `33332784224`, Docker `33332783891`, and Nix `33332783919`; `ad949774...` is green in CI `33350873069`, Docker `33350872546`, and Nix `33350872477`. Live main is still exactly two Photon/docs commits ahead of the submitted base, and that interval is path-disjoint from all eight files in this PR.

The remaining acceptance blocker from my prior review is still present and this moved head does not address it: the submitted diff still modifies `hermes_cli/web_server.py` around the 3,500-line region. The extraction shrinks that godfile by ~53 lines, but a touched production file that remains above the repository's 2,000-physical-line ceiling is not accepted under the current structural gate. The second commit changes only `hosted_rooms.py` and its tests, so it cannot close that blocker.

Required closure is unchanged: either land the install-identity extraction through its owning godfile-sharding path and rebase this room slice so `hermes_cli/web_server.py` is no longer part of this PR's diff, or complete enough of the owning extraction here to leave the touched file under 2,000 lines. Then recompose the final carrier on landing `main` and reacquire CI/Docker/Nix on every surviving commit.

I do not see a new behavioral blocker introduced by `ad949774...`. The authority/replay mechanics and the added lineage proof are coherent; the remaining blocker is structural acceptance plus final current-main rematerialization.

teknium1:
Merged via #99007 (rebase-merge, your commit and authorship preserved) — https://github.com/NousResearch/hermes-agent/pull/99007, on main at cc4b5ba1fc.

The decision you asked for on #97681 is confirmed: root `state.db` behind the `groups.*` surface is the authority and replay boundary. Two follow-ups rode on top in the salvage PR:

- @andrexibiza's blocker 1 (UTF-8 replay bound) is pinned by a discriminating regression test — the current head's byte accounting was already correct, but the prior multibyte test passed even under char-based accounting; the new test fails under it.
- Per the durability requirement Teknium added on #97681, `read_events()` pages now carry the room's `authority.gateway_id` + `authority.epoch` stamp so participant gateways can persist lineage with every replicated page — groundwork for continuous log replication + epoch takeover.

Next up the maintainer side: #97744, then #97797/#97846/#98072 with their owning blockers fixed during salvage. Excellent layering discipline on this stack — thank you.

PR #97715, opened 2026-08-29
fix(skills): keep the skill and snapshots when a batch rollback restore fails

### LINKED ISSUE #97714: [Bug]: a failed batch rollback restore deletes the skill and its snapshots
### Bug Description

The `operations[]` batch in `skill_manage` snapshots every touched skill, then rolls back on failure. The rollback step removes the live skill directory FIRST and copies the snapshot back SECOND:

```python
if post_dir is not None and post_dir.is_dir():
    shutil.rmtree(post_dir)
shutil.copytree(snap, pre_dir)
```

If the copy fails (disk full, locked file, path too long on Windows), the except only appends a `ROLLBACK FAILED` note, and the `finally` then runs `shutil.rmtree(snap_root, ignore_errors=True)`, deleting the only remaining copy. The batch returns a failure payload, but the skill directory and its snapshot are both gone. A feature whose contract is atomicity destroys the skill entirely on a rollback error.

### Steps to Reproduce

1. Create a skill, then run a two op batch where the second op fails (for example `write_file` to an invalid path).
2. Make the restore `copytree` raise (for example a monkeypatched `shutil.copytree` failing on its second call).
3. On current main the skill directory does not exist afterwards and the snapshot directory is deleted too.

### Expected Behavior

A failed restore never destroys data. The skill survives in its half applied state, and the snapshots stay on disk so the operator can recover by hand.

### Actual Behavior

The skill directory is removed before the restore lands, the restore fails, and the finally deletes the snapshots. Nothing survives.

### Affected Component

N/A (CLI only)

### Debug Report

```shell
N/A. Found by reading #97295 against current main. Reproduced with a unit test (test fails on main: the SKILL.md path does not exist after the failed rollback).
```

### Operating System

Windows 11

### Python Version

3.11

### Hermes Version

main 24e54b55f

`tools/skill_manager_tool.py` `_rollback()` deletes `post_dir` before `copytree(snap, pre_dir)`. The destructive order means any restore failure loses both copies, because the caller's `finally` unconditionally removes `snap_root`.

### Proposed Fix (optional)

Rename the broken state aside, restore the snapshot, and delete the aside only after the restore succeeds. If the restore still fails, rename the broken state back. When rollback reports any failure, keep `snap_root` and log its path instead of deleting it in the `finally`.

PR will follow this issue immediately. Follow-up to #97692, same executor.

Adolanium:
## What does this PR do?

Follow-up to #97692, same batch executor. Rollback used to `rmtree` the live skill directory before `copytree` restored the snapshot. When the copy failed (disk full, locked file, path too long on Windows) the except only added a note, and the `finally` then deleted the snapshot directory too. Nothing survived, and the batch still returned a normal failure payload.

The broken state is now moved to a private temp dir first and deleted only after the snapshot is restored. That temp dir is not under `skills/`, so it cannot collide with a real skill name (the old sibling name `skill.rollback-broken` is a valid skill name). If the restore still fails, the broken state is moved back, so the worst outcome is the half applied batch instead of no skill at all. When rollback reports any failure the snapshots are kept on disk and their location is logged at warning level, instead of being deleted by the `finally`.

## Related Issue

Fixes #97714

Follow-up to #97692.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes a bug)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `tools/skill_manager_tool.py`: `_rollback()` restores swap-safe (move aside into a private temp dir, copy, delete aside, move back on failure). The `finally` keeps `snap_root` and logs its path when rollback reported any failure.
- `tests/tools/test_skill_manage_batch.py`: restore `copytree` is forced to fail. The half applied skill survives, a sibling named `probe.rollback-broken` is left alone, and the pre-batch snapshot is kept.

## How to Test

1. `scripts/run_tests.sh tests/tools/test_skill_manage_batch.py -q`
2. New test `test_failed_restore_never_destroys_the_skill` fails on current main (`SKILL.md` does not exist after the failed rollback). With the fix the skill survives with the half applied patch, the error carries `ROLLBACK FAILED`, the snapshots are kept, and `probe.rollback-broken` is not deleted.
3. Regression check: `scripts/run_tests.sh tests/tools/test_skill_manager_tool.py tests/tools/test_skill_manage_schema_diet.py`

Ran on Windows 11 with Python 3.11: 63 tests passed across the three skill_manage test files.

### Code

## For New Skills

N/A

### DIFF
--- a/tests/tools/test_skill_manage_batch.py
+++ b/tests/tools/test_skill_manage_batch.py
@@ -176,6 +176,42 @@ class TestSkillManageBatch(unittest.TestCase):
         self.assertNotIn("Step A.", content)
         self.assertFalse(os.path.exists(os.path.join(self.home, "skills", "beta")))
 
+    def test_failed_restore_never_destroys_the_skill(self):
+        """Rollback used to rmtree the live skill directory BEFORE
+        copytree restored the snapshot. When copytree failed (disk full,
+        locked file) the except only appended a note and the finally then
+        deleted the snapshot too: nothing survived. The broken state must
+        be moved aside and only deleted once the restore succeeded."""
+        from unittest.mock import patch as _patch
+
+        import shutil as _shutil
+
+        self._call("probe", [{"action": "create", "content": SK.format(n="probe")}])
+        state = {"n": 0}
+        real_copytree = _shutil.copytree
+
+        def flaky_copytree(src, dst, *a, **k):
+            state["n"] += 1
+            if state["n"] == 2:  # call 1 snapshots, call 2 is the restore
+                raise OSError("disk full")
+            return real_copytree(src, dst, *a, **k)
+
+        with _patch("shutil.copytree", side_effect=flaky_copytree):
+            r = self._call("probe", [
+                {"action": "patch",
+                 "old_string": "Step 1.", "new_string": "Step ONE."},
+                {"action": "write_file",
+                 "file_path": "bad/nope.md", "file_content": "x"},
+            ])
+        self.assertFalse(r["success"], r)
+        self.assertIn("ROLLBACK FAILED", r["error"])
+        # The skill directory was NOT destroyed by the failed rollback:
+        # the half applied state survives instead of nothing at all.
+        skill_md = os.path.join(self.home, "skills", "probe", "SKILL.md")
+        self.assertTrue(os.path.exists(skill_md))
+        content = open(skill_md).read()
+        self.assertIn("Step ONE.", content)
+
     def test_single_op_path_unchanged(self):
         self._call("probe", [{"action": "create", "content": SK.format(n="probe")}])
         raw = self.smt.skill_manage(
--- a/tests/tools/test_skill_manage_batch.py
+++ b/tests/tools/test_skill_manage_batch.py
@@ -22,6 +22,7 @@ SK = (
 class TestSkillManageBatch(unittest.TestCase):
     def setUp(self):
         self.home = tempfile.mkdtemp(prefix="skmbatch_t_")
+        self._keep_dirs = []
         os.environ["HERMES_HOME"] = self.home
         os.environ["HERMES_YOLO_MODE"] = "1"
         os.makedirs(os.path.join(self.home, "skills"), exist_ok=True)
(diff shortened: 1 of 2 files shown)


lEWFkRAD:
**Blocking review (formal CHANGES_REQUESTED is unavailable to this account):** the rollback destination is still not collision-safe.

1. **P1 - unrelated data can be deleted.** `aside = post_dir.with_name(post_dir.name + '.rollback-broken')` is a predictable sibling name, and `.rollback-broken` is permitted by the existing skill/category name grammar. `shutil.rmtree(aside, ignore_errors=True)` therefore deletes a legitimate pre-existing sibling skill/category before the rollback starts. Please allocate an exclusive temporary path (for example with `tempfile` under a private rollback root), never a user-addressable skill name, and fail closed if an exclusive move cannot be established.

2. **Cleanup failure is reported as success.** After a successful `copytree`, `shutil.rmtree(aside, ignore_errors=True)` can silently leave the half-applied directory behind. Because that sibling name is itself discoverable as a valid skill name, this can leave a stale duplicate active while `_rollback()` reports success and the `finally` removes the recovery snapshot. Cleanup failure must be surfaced and the snapshot retained until the aside copy is positively removed (or moved to a non-discoverable private location).

3. **The regression test does not prove the recovery invariant.** It only asserts that the half-applied live skill survives. It never asserts that the retained snapshot exists, is the exact pre-batch tree/content, and is uniquely attributable to this operation; the failed-rollback path also leaks its `skill_batch_*` temp tree. Please bind the temp location to the test, assert the exact original snapshot is retained (without scanning unrelated global temp directories), and clean it up explicitly.

These are blocking because the proposed recovery path can itself erase unrelated valid user data and can falsely report a complete rollback.

kshitijk4poor:
Merged via salvage PR #97924 (rebase merge — your commit landed on main as part of 154fd10af0's history with your authorship preserved). Your fix won the duplicate-pair evaluation against #97748: the move-aside restore preserves a working half-applied tree when the restore itself fails, which is strictly better recovery than staging-then-swap. #97748's snapshot-path error detail was folded in on top with credit.

https://github.com/NousResearch/hermes-agent/pull/97924

Thanks for the fix!

PR #97717, opened 2026-08-29
fix(cli): resume accepts unambiguous session ID prefixes

### LINKED ISSUE #97716: [Feature]: let --resume accept session ID prefixes like export/delete do
### Problem or Use Case

**Premise corrected** (see the comment below, kept as the record): the original filing claimed `hermes sessions list` prints truncated IDs that cannot be pasted back into `--resume`. That is wrong on current main. `sessions list` prints full IDs, the `/resume` hint prints full IDs, and the `[:18]` truncation that does exist sits in the interactive curses picker, which is arrow-key driven and returns the full internal ID on selection.

What remains: `--resume` and `-c` accept exact IDs and titles only, while every other ID-accepting surface already resolves through `SessionDB.resolve_session_id` (exact or unambiguous prefix):

- `hermes sessions export` / `delete` (`hermes_cli/sessions_cmd.py`)
- the gateway (`gateway/run.py`, `/resume` dispatch)
- the console engine (`hermes_cli/console_engine.py`)

The plain resume path looks like a surface missed when that unification landed, rather than a deliberate exact-match choice.

### Proposed Solution

Route `_resolve_session_by_name_or_id` in `hermes_cli/main.py` through `SessionDB.resolve_session_id`, the same call the three surfaces above already use. Ambiguous prefixes keep falling through to title resolution, and the compression tip projection is unchanged.

### Alternatives Considered

Keep resume exact-match-only for strictness. The resolver is already the established pattern everywhere else in the CLI, and an ambiguous prefix resolves to nothing rather than a wrong session, so strictness is not lost.

Adolanium:
Allow CLI resume to resolve an unambiguous session-ID prefix through SessionDB.resolve_session_id, matching other session commands. Exact IDs and title fallback remain supported, and compressed sessions still resolve to the current continuation. Tests use a real temporary session store.

Fixes #97716

Validation: The new prefix regression failed on main and passed with the fix. Across five resume test files, 42 tests passed; test_in_dir_expands_user_home fails identically on unchanged main on Windows because its HOME-only fixture does not redirect Windows home expansion.


Adolanium:
Correction posted on #97716 and the PR body rewritten to match. Short version: the truncation premise was wrong for current main, sessions list prints full IDs and the [:18] picker is interactive so its truncation never reaches a paste workflow. The remaining value of this diff is consistency, resume now resolves exact-or-prefix like sessions export/delete, the gateway, and the console engine already do. Type flipped from bug fix to feature. Happy to withdraw if exact-match-only was intended.

Adolanium:
I updated this on the refactored main and pushed d27ce22ca8 as one commit. I applied the prefix lookup to the current resolver and preserved compression-tip handling. I also moved the tests to tests/hermes_cli and replaced unmanaged temporary state with isolated real-database fixtures.

The new prefix regression failed on main and passed with the fix. Across five resume test files, 42 tests passed; test_in_dir_expands_user_home fails identically on unchanged main on Windows because its HOME-only fixture does not redirect Windows home expansion.

PR #97721, opened 2026-08-29
fix(read_file): scanned PDF bytes get the hosted OCR path too

### LINKED ISSUE #97720: [Bug]: scanned PDFs read through a remote backend get no hosted OCR and no NEEDS-OCR warning
### Bug Description

`read_file` has two doors into the bundled anydoc converter:

- the path door (`_extract_anydoc`), used for local files. On `NeedsOcrError` it attempts hosted OCR when a route exists and otherwise returns the NEEDS-OCR warning with the page list and recovery commands.
- the bytes door (`_extract_anydoc_bytes`), used when the document is transferred from a remote terminal backend (`read_file_bytes` + `extract_document_bytes` in `tools/file_tools.py`).

The bytes door has no `NeedsOcrError` handling at all. The same scanned PDF that gets OCR (or a teaching warning with recovery steps) when read locally produces a bare `NeedsOcrError: ...` convert error when read over SSH or any remote backend: no OCR attempt, no page list, no recovery text.

`to_markdown_bytes` accepts the same `ocr`, `api_key`, and `api_url` kwargs as `to_markdown`, so the asymmetry has no technical reason.

### Steps to Reproduce

1. Point Hermes at a remote terminal backend.
2. Read a scanned PDF (pages with no text layer) with `read_file`.
3. The result is a generic convert error. The same file read from a local backend gets hosted OCR or the NEEDS-OCR warning.

### Expected Behavior

Both doors behave identically for the same document.

### Actual Behavior

Only the path door handles scanned pages.

### Affected Component

N/A (all platforms)

### Debug Report

```shell
N/A. Found by reading #97195 against current main. Reproduced with unit tests (three new cases for the bytes door fail on main with ExtractionError instead of OCR text or the warning).
```

### Operating System

Windows 11

### Python Version

3.11

### Hermes Version

main 24e54b55f

`tools/read_extract.py` `_extract_anydoc_bytes` catches every exception as a generic `ExtractionError` and never checks the typed `NeedsOcrError` signal, unlike `_extract_anydoc` a few lines above.

### Proposed Fix (optional)

Mirror the path door: catch `NeedsOcrError`, attempt `to_markdown_bytes(data, ocr='hosted', ...)` through the same `_hosted_ocr_config` gate, and return `_needs_ocr_warning(...)` when no route exists or the attempt fails.

PR will follow this issue immediately. Follow-up to #97688, same gate and same warning.

Adolanium:
Handle scanned PDFs received as bytes through the existing _ocr_scanned_pdf recovery helper. Both SDK entry points use explicit credentials and endpoint arguments. Successful hosted OCR returns normalized text without warning about the original PDF's empty native text layer; disabled or failed OCR returns the usual recovery guidance.

Fixes #97720

Validation: 78 tests passed via scripts/run_tests.sh across extraction, schema gating, and UTF-8/binary regression tests. The real anydoc SDK test intercepts the HTTP request and verifies that bytes OCR cannot inherit process credentials or the endpoint from another profile.


Enough1122:

Reviewed `tools/read_extract.py` + tests for PR #97721. Follow-up to #97688: the bytes door (`_extract_anydoc_bytes`, used for backend-transferred documents) now handles `NeedsOcrError` exactly like the path door — hosted OCR attempt when a route exists, NEEDS-OCR teaching warning otherwise. The `_finish_anydoc_bytes` extraction removes real duplication (empty check, newline trim, prepended PDF coverage note) and preserves prior behavior byte-for-byte. Tests cover hosted success, disabled (no attempt), and hosted-failure (named error) cases.

Non-blocking observations:

1. **Hosted failure collapses into the NEEDS-OCR warning.** The broad `except Exception` around the hosted attempt converts any hosted error (HTTP 500, 401, timeouts) into a `[NEEDS OCR] … attempted and failed` teaching warning rather than a surfaced error. This mirrors the path door's contract, so it's consistent — but the recovery guidance is aimed at a *scanned-PDF* cause, which may be misleading when the real cause is a broken hosted route. The warning does name the underlying error, which mitigates this.

2. **Test scaffolding indirection.** `_fake_mod` building a module object with a dynamically-typed `NeedsOcrError` is a fair amount of machinery, and `patch.object(rx, "_anydoc", ...)` with inline-whitespace line breaks reads awkwardly. Cosmetic; the assertions themselves are meaningful.

3. **Coverage-note prepend preserved** — `_finish_anydoc_bytes` reuses `_pdf_coverage_note_from_bytes` and prepends (correct for paginated reads); no regression observed in that path.

The change closes a real parity gap (remote-backend scanned PDFs previously got a bare convert error) and keeps the two doors consistent. Nothing blocking.

Adolanium:
I updated this on the refactored main and pushed 309aa66ac5 as one commit. I ported the bytes recovery into the shared helper and fixed the misleading native-scan warning after successful OCR. Both SDK calls now preserve the explicit credential handling needed by #97688.

78 tests passed via scripts/run_tests.sh across extraction, schema gating, and UTF-8/binary regression tests. The real anydoc SDK test intercepts the HTTP request and verifies that bytes OCR cannot inherit process credentials or the endpoint from another profile.

PR #97723, opened 2026-08-29
fix(gateway): cap concurrent handoff dispatches in the watcher

### LINKED ISSUE #97722: [Bug]: a backlog of pending handoffs opens one parallel agent turn per row
### Bug Description

The `/handoff` watcher dispatches claimed rows fire-and-forget (`inflight[session_id] = asyncio.ensure_future(_dispatch(...))` in `gateway/run.py`). Each dispatch runs a FULL agent turn plus delivery. There is no cap on how many run at once.

After an outage, or whenever several sessions hand off close together, the next poll claims every pending row and opens N parallel model calls in the same instant. delegation caps its concurrent children (`delegation.max_concurrent_children`, default 3) for exactly this class of work, the handoff path has nothing.

### Steps to Reproduce

1. Queue several `/handoff` requests while the gateway is down (or mock `list_pending_handoffs` to return 5 rows).
2. Start the gateway. The first tick claims all 5 and starts 5 concurrent full agent turns.

### Expected Behavior

Concurrent handoff dispatches are capped. Extra rows stay queued (claimed, deduped by the inflight map) until a slot frees.

### Actual Behavior

All claimed rows dispatch simultaneously with no limit.

### Affected Component

- Gateway (Telegram/Discord/Slack/WhatsApp)

All platforms through the handoff path

### Debug Report

```shell
N/A. Found by reading #97338 against current main. Reproduced with a unit test (5 pending rows, max observed concurrency is 5 on main).
```

### Operating System

Windows 11

### Python Version

3.11

### Hermes Version

main 24e54b55f

`gateway/run.py` `_handoff_watcher`: `_tick` creates one task per claimed row with no semaphore, matching `delegation`'s uncapped opposite.

### Proposed Fix (optional)

A closure-local `asyncio.Semaphore(_HANDOFF_MAX_CONCURRENT)` (3, matching delegation's default) held for the agent turn inside `_dispatch`. Terminal-state bookkeeping stays outside the slot so completion never queues behind the cap.

PR will follow this issue immediately. Follow-up to #97694, same watcher.

Adolanium:
Bound concurrent handoff agent turns in gateway/run_adapters.py. The watcher continues polling and claiming across profiles while excess dispatches wait for a semaphore slot. Completion bookkeeping remains outside the cap.

Fixes #97722

Validation: 23 tests passed via scripts/run_tests.sh across all five gateway handoff test files. The backlog regression checks configured-cap concurrency and completion of every row.


Enough1122:

Reviewed `gateway/run.py` + tests for PR #97723. Adds `_HANDOFF_MAX_CONCURRENT = 3` with a per-watcher `asyncio.Semaphore` so a post-outage backlog of N pending handoff rows cannot open N parallel full agent turns. The design is well-chosen: only the agent-turn portion of the dispatch holds a slot, while `complete_handoff` terminal-state bookkeeping runs outside the semaphore, so completion never waits on the cap. The regression test is mutation-survivable (it would catch removal of the cap) and verifies all queued rows eventually dispatch.

Non-blocking observations:

1. **The cap is per-watcher-instance, shared across all watched scopes.** If `_handoff_watch_scopes` returns multiple profiles, they share the single 3-slot pool. Probably desired (it's a global model-call cap), but worth confirming that's the intent rather than per-profile.

2. **Cancellation path releases the slot correctly** — `async with _dispatch_slots` releases on `CancelledError` propagation through the context manager, and the watcher's `except asyncio.CancelledError` handles gateway shutdown. Verified from the surrounding code; no leak.

3. **Minor:** the closure-local `_dispatch_slots` exists specifically to avoid hanging state on `self` for test stand-ins — pragmatic, but it also means the cap resets on every watcher restart rather than being a persistent property of the runner. For a fire-and-forget watcher that's fine; just noting the semantics.

The change caps a real resource-exhaustion class and matches delegation's existing default of 3. Nothing blocking.

Adolanium:
I updated this on the refactored main and pushed 54457a0b16 as one commit. I moved the cap into the refactored adapter lifecycle module and strengthened the regression to check that every queued handoff completes.

23 tests passed via scripts/run_tests.sh across all five gateway handoff test files. The backlog regression checks configured-cap concurrency and completion of every row.

Issue #89886, opened 2026-08-19
v2026.8.18: cache_control on tool_result.content[] rejected by Anthropic-format API (non-retryable 400, kills any tool-using session)


johmobit:
## Summary

On `v2026.8.18`, the very first LLM turn of any session that includes a tool result fails with a non-retryable HTTP 400 from the upstream (Anthropic-format) API:

```
litellm.BadRequestError: OpenAIException - Error from provider (Console): Upstream request failed:
[invalid_request_error] messages.2.content.0.tool_result.content.0.cache_control: cache_control may
not be specified within `tool_result.content`. Instead, place it directly on `tool_result`. Received
```

The agent appears to be placing a prompt-cache `cache_control` breakpoint inside a `tool_result` message's `content[0]`, rather than on the `tool_result` block itself. The upstream API rejects this outright, and since it's classified as a non-retryable client error, the whole job/session aborts immediately — no retry, no fallback, no partial output.

## Impact

Any cron/scheduled job (in our case, a daily-briefing cron task using MCP tool calls) fails on its first turn with zero output. Concretely: `cron.scheduler: Job 'daily-briefing' failed: RuntimeError: HTTP 400: ...`, leaving behind a session with no assistant content at all.

## Environment

- Image: `nousresearch/hermes-agent:v2026.8.18`
- LLM gateway: LiteLLM (OpenAI-compatible proxy) in front of an Anthropic-format ("Console") provider
- Trigger: any session where a tool call precedes a cached turn (we saw it on the very first cron-triggered LLM call, which followed an MCP tool invocation)

## Regression

This did not occur on the previous pinned version, `v2026.7.30`. We've rolled back to that tag as a workaround. Not yet confirmed whether this is new agent-side request-construction logic in `v2026.8.18` or a change in how it's exercised — happy to help narrow it down further if useful (e.g. a minimal repro session), but wanted to file this now since it makes any tool-using scheduled job fail outright with no output.

liuhao1024:
I traced the request-construction chain on current main; the mechanism is exactly what the error text says, and the injection point is identifiable:

**Where the marker lands inside `tool_result.content`.** `agent/prompt_caching.py:_apply_cache_marker` handles `role:"tool"` messages on the non-native (envelope/OpenAI-format) layout by converting string content to a part list and putting `cache_control` **on the content part**:

```python
msg["content"] = [
    {"type": "text", "text": content, "cache_control": cache_marker}
]
```

[...]

PR #87782, opened 2026-08-16
fix(computer_use): stop approval failing open when no CLI callback is wired

jackulau:
## What does this PR do?

`computer_use`'s destructive-action gate failed open whenever no CLI approval callback was registered:

```python
    cb = _approval_callback
    if cb is None:
        # No CLI approval wired — default allow. Gateway approval is handled
        # one layer out via the normal tool-approval infra.
        return None
```

The second sentence is what makes this safe, and it is not true by default. The chain, checked against `56526bc0d`:

1. The generic per-tool human gate is `tools.approval.request_tool_approval`. Its only non-test caller is `hermes_cli/plugins.py:2522`, reached when a plugin `pre_tool_call` hook returns `{"action": "approve"}`.
2. `ToolRegistry.dispatch` calls `entry.handler(args, **kwargs)` with no approval step of its own.
3. `model_tools.py`'s pre-dispatch path gates ACP *edit* approval (`maybe_require_edit_approval`, file mutations only) and sets approval observability context. There is no tool-name-keyed check for `computer_use`.
4. `tools/approval.py`'s own gate is command-shaped (`detect_hardline_command`, `check_dangerous_command`, `_SENSITIVE_WRITE_TARGET`). `computer_use` never presents a command string to it.

So the "one layer out" protection exists, but nothing routes `computer_use` into it. An operator who installed a plugin rule naming the tool is covered; a default install is not. Every non-CLI entry point (gateway, cron, `-q`, a bare script) could click, drag, type and press keys on a real desktop with nobody asked.

**The `bring_to_front` rung goes with it, and that is the part worth a second look.** #67052 deliberately keeps persistent focus in a *separate* approval scope from the input itself, on the reasoning that a background approval must not silently authorize a visible focus change. Both rungs call `_request_approval`, so with no callback both returned `None`: the distinction that issue paid for was absent precisely where nobody is watching the screen.

### The fix

The `cb is None` branch now calls `request_tool_approval` instead of returning `None`. This makes the existing comment true rather than replacing it with a new policy, and `computer_use` inherits the contexts that function already handles:

| context | before | after |
|---|---|---|
| interactive CLI (callback registered) | callback decides | unchanged, this branch is not reached |
| `--yolo` | allowed | allowed (the shared gate bypasses first) |
| gateway | **allowed silently** | notify round-trip, or a pending approval for `/approve` |
| cron | **allowed silently** | honors `approvals.cron_mode` |
[...]

PR #87764, opened 2026-08-16
fix(tools): fail closed on computer_use approval when no callback is wired

QDung210:
## What does this PR do?

Fixes a fail-open approval default in `computer_use`. `_request_approval()` in `tools/computer_use/tool.py` returned `None` (allow) unconditionally whenever no CLI approval callback was wired:

```python
cb = _approval_callback
if cb is None:
    # No CLI approval wired — default allow. Gateway approval is handled
    # one layer out via the normal tool-approval infra.
    return None
```

The interactive CLI (`cli.py`) is the *only* call site that ever wires a real callback via `set_approval_callback`. Cron jobs, every bot-platform gateway (`computer_use` lives in `_HERMES_CORE_TOOLS`, shared by `hermes-cron` and the messaging toolsets per `toolsets.py`), and ACP sessions never call it, so any of those contexts could execute destructive actions (click, type, drag, key combos) with zero confirmation. The comment's claim of an "outer gateway layer" handling approval isn't evidenced anywhere: `tools/registry.py`'s `dispatch()` has no generic approval gate, and `tools/approval.py` (the real dangerous-command-detection system) never references `computer_use`.

Reproduced live against the test noop backend: with `_approval_callback` at its default `None`, a `click` action executed successfully — no prompt, no error.

**Fix:** when `cb is None`, check `is_approval_bypass_active_for_session()` across both session-identity namespaces (DB `session_id` and gateway `session_key`) — the exact same bypass check `_cua_permission_mode()` already performs a few lines above in the same file, for the same reason. If the user has explicitly opted into unattended operation (`/yolo`, `HERMES_YOLO_MODE`, or `approvals.mode: off`), the action still proceeds exactly as before. Otherwise it now fails closed with an explicit error instead of silently allowing.

Read-only actions (`_SAFE_ACTIONS`: capture, wait, list_apps, list_windows, cua_browser_state) never reach `_request_approval` at all and are unaffected. A wired CLI callback still takes precedence over the bypass check (covered by a new test) — this only changes the no-callback fallback.

**Note for maintainers:** open PR #85075 ("honor approval bypass mode") touches the same function for a different gap (making bypass suppress even a *wired* callback). The two changes are logically compatible but will likely need a small rebase against each other depending on merge order.

## Related Issue

Fixes #87724

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [x] 🔒 Security fix
[...]

Issue #87724, opened 2026-08-16
computer_use mutations fail open in headless dispatch without an approval callback


tachyon-r:
## Bug description

`computer_use` classifies user-visible mutations as approval-gated, but the handler explicitly allows them when the interactive CLI callback is absent. Headless registry/API execution therefore reaches the backend without an approval responder or decision.

## Code evidence

- Mutating actions are listed in `_DESTRUCTIVE_ACTIONS`: `tools/computer_use/tool.py:80-92`.
- The gate runs for those actions at `tools/computer_use/tool.py:515-528`.
- `_request_approval()` returns approved when `_approval_callback is None`: `tools/computer_use/tool.py:570-574`.
- The callback is installed by the interactive CLI: `cli.py:7898-7907`.
- Registry dispatch invokes the handler directly: `tools/registry.py:1102-1127`.

## Deterministic probe

With the noop backend and no callback:

```python
cu.set_approval_callback(None)
result = cu.handle_computer_use(
    {"action": "click", "coordinate": [10, 20]},
    session_id="headless-audit",
)
```

Observed:

```text
result={"ok":true,"action":"click",...}
backend_calls=[('click', {'x':10,'y':20,...})]
```

Focused probes also returned approved for `type`, foreground `focus_app`, and `cua_browser_prepare(profile_mode="existing_profile")`.

## Expected behavior

A mutation with no reachable approval responder fails closed, or is allowed only by an explicit immutable unattended/YOLO capability.

## Actual behavior

Absence of the CLI callback is treated as approval.

## Impact

Headless gateway, API, cron, ACP, or worker dispatch can mutate user-visible desktop/browser state without a human decision. Existing-profile browser preparation is included in the fail-open set.

## Suggested fix

Pass an explicit approval capability into the handler. Default to deny when no responder/capability exists; do not infer approval from callback absence. Add registry-level tests for every headless surface, not only direct CLI callback tests.

## Environment

- Hermes Agent v0.20.1
- Audited commit: `56526bc0d36522ab7a87ee0056f70e3847d2f0e6`
- Reproduced with noop backend; no real desktop mutation required

jackulau:
Confirmed against `56526bc0d`, and I went looking for the reason to dismiss this before accepting it, because the fail-open is not an oversight. It is annotated:

```python
    cb = _approval_callback
    if cb is None:
        # No CLI approval wired — default allow. Gateway approval is handled
        # one layer out via the normal tool-approval infra.
        return None
[...]

PR #97724, opened 2026-08-29
fix(computer-use): fail closed when no approval prompt surface is wired

pixelatedhacker:
## What does this PR do?

Closes the fail-open approval default in `computer_use`'s `_request_approval()` (#87724): when no approval prompt is wired for the current execution context, destructive actions (click, type, drag, key, …) used to run with zero confirmation — the comment claimed an outer gateway layer handled approval, but `tools/approval.py` never referenced `computer_use` and no such layer exists.

With no wired callback, destructive actions now escalate through the repository's normal human-approval gate (`tools.approval.request_tool_approval`):

- gateway sessions get the standard approval round-trip (embed + buttons, timeout fail-closed);
- cron jobs honor `approvals.cron_mode` (default deny);
- single-query (`-q`) sessions honor `approvals.single_query_mode`;
- a context with no possible responder fails closed with `code: approval_unavailable` — silence is not consent.

Grant scoping: the gate's `rule_key` embeds the approving session key (`computer_use:<session_key>:<cu_session_id>:<action>:<delivery>`), so an "Always" answered in an interactive session is scoped to that session and cannot authorize a cron/single-query/headless context that reuses the same computer_use `session_id`. This deliberately sidesteps the shared-gate ordering defect tracked in #61064 instead of widening it; computer_use's own `_always_allow` / `_session_auto_approve` semantics are unchanged.

Callback ownership: `set_approval_callback` stores the callback in a `ContextVar` (per execution context, not process-global) and returns a token for the new `reset_approval_callback()`. A headless context can no longer observe or drive another concurrent context's responder. This composes #81668 inline; if that PR lands first this commit rebases onto it cleanly.

Read-only actions (`capture`, `wait`, `list_apps`, `list_windows`) never reach the gate. The wired interactive CLI path is unchanged.

## Tests

- pristine state (no callback, no responder): destructive action fails closed with `approval_unavailable`, backend untouched (flipped isolation regression)
- concurrency: a callback-free context never invokes a concurrent context's responder and fails closed
- prior interactive grant → cron run still denied; prior interactive grant → single-query run still denied
- positive controls: a grant authorizes its own session
- local: `tests/tools/test_computer_use*.py`, `tests/computer_use/`, approval suites green on this head

Related: #87724 (defect), #87764 (competing carrier — reviewed; this PR takes the shared-gate route with session-scoped grants), #81668 (callback ownership, composed), #61064 (ordering defect, deliberately sidestepped).

Supersedes the earlier hard-deny iteration of this branch after review.

### DIFF
--- a/tests/tools/test_computer_use.py
+++ b/tests/tools/test_computer_use.py
@@ -20,12 +20,17 @@ import pytest
 @pytest.fixture(autouse=True)
 def _reset_backend():
     """Tear down the cached backend between tests."""
-    from tools.computer_use.tool import reset_backend_for_tests
+    from tools.computer_use.tool import reset_backend_for_tests, set_approval_callback
     reset_backend_for_tests()
+    # These tests exercise dispatch and the backend schema, not the approval
+    # prompt. No callback is wired on headless surfaces and _request_approval
+    # fails closed, so grant destructive actions here; conftest clears it.
+    set_approval_callback(lambda action, args, summary: "approve_once")
     # Force the noop backend.
     with patch.dict(os.environ, {"HERMES_COMPUTER_USE_BACKEND": "noop"}, clear=False):
         yield
     reset_backend_for_tests()
+    set_approval_callback(None)
 
 
 @pytest.fixture
--- a/tests/tools/test_computer_use_approval_isolation.py
+++ b/tests/tools/test_computer_use_approval_isolation.py
@@ -55,8 +55,9 @@ def test_a_forgets_a_poisoned_approval_callback():
     def stale_two_arg_callback(action, args):  # wrong arity on purpose
         return "approve_once"
 
-    cu_tool.set_approval_callback(stale_two_arg_callback)
-    # no reset — the autouse fixture must clean this up
+    token = cu_tool.set_approval_callback(stale_two_arg_callback)
+    assert token is not None
+    # deliberately no reset — the autouse fixture must clean this up
 
 
 def test_b_still_fails_closed_without_a_callback():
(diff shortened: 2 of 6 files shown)


alt-glitch:

Related: #87764 addresses the same callback-absent computer_use approval gap with the normal approval-escalation path. This PR instead returns an immediate approval_unavailable denial; these are competing mechanisms, not duplicates.

andrexibiza (commented):
Reviewed exact head `05b9932aeaec6d229b703aec9158317b5ac7426a` against base `10e93c6ab958c7ec61cfc4416f4d4459e72ca8a7` and current `main@1d8946b40b9333a7fda81be890be31771a312d4f`.

The underlying bug is real, and this head does close the **pristine** no-callback fail-open: the new `approval_unavailable` branch prevents a destructive action from reaching the backend when `_approval_callback` is actually `None`, and the isolation regression now pins that state. That is useful work to preserve.

I cannot clear this head yet because two authority boundaries are still wrong.

## P1 — `cb is None` is not equivalent to “no approval surface exists”

The new branch in `tools/computer_use/tool.py::_request_approval()` says:

> `No prompt surface is wired (headless, gateway, messaging runs).`

but the predicate is only `cb = _approval_callback; if cb is None`. `_approval_callback` is the **CLI-local** callback. Current `tools.approval` already owns the wider approval contexts through `request_tool_approval()` / `_run_approval_gate`: gateway sessions can do the normal notification/button round-trip, cron honors `approvals.cron_mode`, single-query honors `approvals.single_query_mode`, and a genuinely non-interactive/no-human context fails closed.

So this head trades the fail-open for an over-broad hard denial:

- a Discord/Telegram/Slack-style gateway user with a reachable approval responder gets `approval_unavailable` instead of the repository's normal approval round-trip;
- `approvals.cron_mode: approve` is ignored for `computer_use`;
- `approvals.single_query_mode: approve` is ignored for `computer_use`.

This is already the key distinction documented on #87724 by jackulau, implemented in closed #87782, and then adopted into QDung210's still-open #87764. Those are not unrelated duplicates: they establish the existing product-policy path this patch currently bypasses.

**Required fix:** when there is no local callback, enter the shared approval authority with the existing `(action, delivery_mode)` scope, and fail closed only when that authority has neither an explicit unattended grant nor a reachable human. Preserve the prior contributors' work rather than re-inventing the policy locally.

There is one merge-order caveat: do not blindly copy #87764's current bridge. Open #61064 tracks the shared-gate ordering defect where a prior `Always` approval can override unattended `cron_mode: deny`. `computer_use` should not introduce a new permanent grant path that widens its existing session-owned `_always_allow` / `_session_auto_approve` semantics. Either compose the shared-gate ordering fix first, or make the bridge explicitly non-permanent/session-owned and add prior-grant → cron-deny + single-query-deny regressions.

## P1 — callback ownership is still process-global, so the headless class is not actually closed

At this head `_approval_callback` is still a module-global value. That means `cb is None` is not a session-owned fact. In a process serving concurrent sessions:

1. session A installs an interactive callback;
2. session B is headless and has no responder of its own;
3. session B observes A's process-global callback, skips this new fail-closed branch entirely, and can route its mutation through A's responder.

That is the other side of the same authority shape: absence/presence must be bound to the owning execution context, not ambient process state. Open #81668 (jjjinz) already isolates this callback with a `ContextVar` and includes a concurrent two-session regression. It is complementary ownership work, not superseded by this PR.

**Required fix:** compose/rebase the session-owned callback isolation (or carry an explicit session-owned approval capability into `handle_computer_use`) and add a concurrency regression where one context has a callback and a second callback-free context performs a destructive action. The second context must never invoke the first context's responder and must enter the shared headless approval path / fail closed.

## Exact-object / interlock state

The PR body says this branch is 499 commits behind tip, but live graph truth is much better: current `main` is only **one** commit beyond this PR's base, and that commit is the disjoint prompt-caching fix #89886. GitHub currently reports the PR mergeable. A current-main refresh should therefore be mechanical after the policy fixes, not a reason to churn the implementation now.

Hosted acceptance is still absent on this exact head: CI run `33238175531`, Docker `33238175259`, and Nix `33238175304` all concluded `action_required`; the CI run has **zero jobs**. The reported local `329` approval/computer-use passes are useful development evidence, but they are not the exact-head hosted release receipt yet.

Ownership map as I see it:

- #87724 — canonical reported fail-open defect (tachyon-r).
- #87782 — closed prior implementation by jackulau that established the shared approval-routing direction; its work was explicitly folded into #87764.
- #87764 — competing current carrier for the same defect; preserve QDung210 + jackulau lineage, but reconcile its permanent-grant/order problem before adopting it wholesale.
- #81668 — complementary callback-ownership isolation required to make “no responder” session-correct.
- #61064 — complementary shared-gate precedence fix for unattended deny policy.
- #97725 — complementary Desktop grant surface; it should consume the same runtime authority/scoping contract, not become a second approval policy.

The minimal deny here is safer than current main's fail-open and the regression is worth keeping. The next step is to make that denial belong to the *right authority*: session-owned callback identity plus the existing gateway/cron/single-query decision path. Once those two boundaries are composed and exact-head CI actually executes, this becomes a much stronger security fix.

pixelatedhacker:
Thanks for the exact-head review — both P1s are fair, and this head now addresses them.

**P1 #1 (cb is None ≠ no approval surface):** agreed. The no-callback branch no longer hard-denies; it escalates through the repository's normal gate via `tools.approval.request_tool_approval()`, so gateway sessions get the standard approval round-trip, cron honors `approvals.cron_mode`, single-query honors `approvals.single_query_mode`, and a context with no possible responder still fails closed (`code: approval_unavailable`). Only the pristine case (`_approval_callback` absent, no responder anywhere) keeps the immediate denial.

**Merge-order caveat:** I did not adopt a permanent-grant bridge. The `rule_key` passed to the gate embeds the *approval* session key (`computer_use:<session_key>:<cu_session_id>:<action>:<delivery>`), so an `[a]lways` answered interactively is scoped to that approval session by construction and cannot short-circuit an unattended context that reuses the same computer_use `session_id`. This sidesteps the shared-gate ordering defect in #61064 rather than widening it; `computer_use`'s own `_always_allow` / `_session_auto_approve` semantics are unchanged. Regressions added: prior interactive grant → cron run still denied (`cron_mode` default deny), and prior grant → single-query run still denied.

**P1 #2 (process-global callback):** `set_approval_callback` now stores the callback in a `ContextVar` and returns a token for a new `reset_approval_callback()` (same shape as #81668 — this composes that ownership fix inline; happy to rebase onto #81668 if it lands first, the two diffs touch the same seams). The concurrency regression you asked for is in: a callback-free context performing a destructive action while another context's responder is installed must never invoke that responder — it enters the shared headless approval path and fails closed.

**Base/CI:** rebased onto current `main` (1d8946b40b93...). Hosted CI still shows `action_required` on my side — I can't self-approve those runs, so maintainer approval is what's left for the exact-head receipt.

pixelatedhacker:
Closing this as a competing duplicate.

Per CONTRIBUTING's guidance on duplicates, #87764 is already open and further along on the same defect (#87724), and the callback-ownership half properly belongs to #81668 — this branch reimplemented both locally instead of building on them.

@andrexibiza thank you for the detailed exact-head review. The one finding worth carrying forward I've left on #81668: the `ContextVar` approach that both that PR and this one adopted looks like it breaks the interactive CLI path, because the callback is installed on the main thread (`cli.py:8430`) while the agent runs in a separate daemon thread (`cli.py:16838`) that starts with an empty context. Neither PR's tests would catch it.

PR #77666, opened 2026-08-03
fix(terminal): detect bind mounts in Docker --mount syntax

Baophan00:
_docker_volume_uses_host_path only flags short-form bind mounts (/host:/container, ./src:/app). Docker --mount syntax (type=bind,source=/host,target=/container) was never detected, so a sandbox declared with long-form syntax silently gained host access. Added --mount parsing that extracts the source= field and checks whether it starts with a host-path prefix.

PR #42750, opened 2026-06-09
fix(security): remove container command guard bypass for defense-in-depth

LifeJiggy:
Removes the container environment bypass in `check_dangerous_command()` and `check_all_command_guards()` so hardline and dangerous command detection applies to all environments (docker, singularity, modal, daytona, vercel_sandbox).

**Changes:**
- Removed `if env_type in {docker, singularity, modal, daytona}: return {approved: True}` from both guard functions in `tools/approval.py`
- Updated 3 test files to assert containers now enforce hardline blocks

Part of #24942

PR #78035, opened 2026-08-03
feat(terminal): add native Apple Container backend for macOS

ajzrva-sys:
## Summary

Salvages and completes #6867 on current `main`, preserving the original contributor's commits and authorship while addressing the sweeper review and validating the backend against Apple Container 1.2.0 on macOS 26 arm64.

- Adds a native Apple Container terminal backend for macOS 26+ on Apple Silicon.
- Wires backend configuration through terminal, file, `execute_code`, prompt/probe, setup, doctor, web API, and Desktop settings surfaces.
- Preserves current shared configuration bridges, including Docker network and Vercel behavior.
- Applies remote-backend cwd sanitization, prompt host-information suppression, host-mount approval routing, and sensitive-volume warnings.
- Supports persistent `/workspace` and `/root`, graceful cleanup, stdin, resource limits, custom images, and user volumes.
- Stages credential files into least-privilege read-only directory mounts because Apple Container rejects regular-file bind mounts; nested skills/cache/user mounts remain supported.
- Uses Apple Container's path-only `--tmpfs` syntax so scratch directories are actually writable over the read-only root filesystem.

## Provenance and related work

- **Salvages #6867:** the original feature and review-fix commits are retained with their original authorship.
- **Addresses the #6867 sweeper review:** configuration now reaches every tool factory; remote-backend contracts and prompt behavior are covered; credential, skills, and cache mounts are read-only.
- **Related alternate implementation: #57127.** This branch was rebuilt on current `main` and avoids the Docker network/config and host-access regressions identified in that PR's review.

## Native validation

Tested with Apple Container CLI 1.2.0 on macOS 26 arm64:

```text
HERMES_RUN_APPLE_CONTAINER_INTEGRATION=1 \
  venv/bin/python -m pytest -q -s \
  tests/integration/test_apple_container_terminal.py

1 passed
```

The native integration test verifies:

- Linux arm64 execution through the terminal backend
- shared terminal/file/`execute_code` lifecycle
- writable workspace and scratch filesystems
- read-only credential exposure with an unchanged host SHA-256
- container removal after cleanup
- a second lifecycle using the same persistent task directories

## Additional verification

- Focused Python regression suite: `735 passed, 2 skipped`
- Apple Container environment tests: `38 passed`
- Desktop Vitest: `38 passed`
- Ruff: passed
- ESLint: passed
- `git diff --check`: passed
- Independent blocking-issue review: passed

## Requirements

- macOS 26 or later
- Apple Silicon
[...]

PR #97727, opened 2026-08-29
fix(terminal): detect host binds declared via docker_extra_args

dewoller:
## What does this PR do?

`_docker_has_host_access()` decides whether a Docker session keeps the approval
fast-path. It reads `docker_mount_cwd_to_workspace` and `docker_volumes` — but not
`docker_extra_args`, which is forwarded to `docker run` verbatim. So `-v /host:/c`
declared there mounts exactly what a `docker_volumes` entry would, while the
function still reports the container as isolated.

Consequence: `_should_skip_container_guards()` short-circuits
`check_dangerous_command()` before its hardline floor, so `rm -rf` against a
bind-mounted host path runs unprompted. Both call sites inherit it — the terminal
guard and the `execute_code` guard, which #54483 wired to the same signal
deliberately.

**This does not propose changing the security model.** It makes the existing one
answer truthfully. On #42750 the maintainer stated the contract directly:

> `tools/approval.py` intentionally retains the fast path for isolated Docker,
> Singularity, Modal, and Daytona backends; Docker exits that path when
> `has_host_access` is true.

A container with a host bind declared via `docker_extra_args` *is* host-exposed
and should exit the fast path. Today it does not. The isolated-container contract
is untouched — an isolated container still takes the fast path.

Per SECURITY.md §2.4 the approval gate is a heuristic, not a boundary, so this is
filed as a normal correctness fix rather than through the private security
channel. It is a config-introspection bug, not a boundary claim.

**Not a deliberate exclusion.** `docker_extra_args` shipped in `ebf2ea584a`
(2026-05-10), seven weeks before the guard in `9860d93f2a` (2026-06-28), whose
message states the general goal "Detect host bind mounts" and then enumerates
only the two mechanisms it covered. `docker_extra_args` is not mentioned anywhere
in that commit, including as an exclusion.

## Relationship to #77666

Adjacent but orthogonal, and worth reading together:

- **#77666** teaches `_docker_volume_uses_host_path()` to parse long-form
  `--mount` syntax inside **`docker_volumes`**.
- **This PR** makes a different config key — **`docker_extra_args`** — get
  inspected at all.

Both touch `--mount` parsing, so they will conflict textually. I am happy to
rebase onto #77666 and reuse its parser rather than carry a second one — say the
word and I will restructure. If #77666 lands first, the `_docker_mount_spec_uses_host_path`
helper here should collapse into theirs.

## Related Issue

No existing issue. Found while auditing a deployment that pins its vault mount
through `docker_volumes` specifically to keep this guard armed.

## Type of Change

## Changes Made

- `tools/terminal_tool.py` — added `_docker_extra_args_use_host_path()` and
  `_docker_mount_spec_uses_host_path()`; `_docker_has_host_access()` now also
  consults `docker_extra_args`.
- `tests/tools/test_modal_sandbox_fixes.py` — added
  `test_docker_extra_args_host_bind_detected` to the existing
  `TestDockerHostBindApproval` class.
- `contributors/emails/dewoller@gmail.com` — attribution mapping for the
  Contributor Attribution Check.

Parses `-v`, `--volume`, `--volume=`, `--mount` and `--mount=`. `--mount` honours
Docker's default of `type=volume`, so only an explicit `type=bind` counts as host
access — a named volume keeps the fast path.

## How to Test

1. Before the change:
   ```python
   from tools.terminal_tool import _docker_has_host_access
   _docker_has_host_access({"env_type": "docker", "docker_volumes": [],
                            "docker_extra_args": ["-v", "/srv/data:/srv/data"]})
   # -> False   (host bind, but reported isolated)
   ```
2. After the change the same call returns `True`, so
   `_should_skip_container_guards("docker", has_host_access=True)` is `False` and
   the command goes through the normal approval flow.
3. `pytest tests/tools/test_modal_sandbox_fixes.py::TestDockerHostBindApproval -q`

Negative coverage included: named volumes, `type=volume`, unrelated flags, a
trailing `-v` with no value, and a non-list config value.

### Code

- [x] I searched for existing PRs to make sure this isn't a duplicate — see the #77666 note above
- [ ] I've run `pytest tests/ -q` and all tests pass — see note below
- [x] I've tested on my platform: macOS 15 (Darwin 25.6), Python 3.11

On the full-suite item, being precise rather than ticking it: I ran
`tests/tools/test_modal_sandbox_fixes.py`, `test_container_cwd_sanitize.py`,
`test_file_tools_container_config.py`, `test_approved_command_clean_slate.py`,
`test_code_execution*.py`, `test_sandbox_failure_hints.py` and `tests/deploy/`
— 181 passed, 2 xfailed. The complete `tests/tools/` run wedged locally on
Docker/network-dependent tests, and it surfaced one failure,
`test_approval.py::TestDetectDangerousRm::test_nonrecursive_verification_artifact_cleanup_is_not_dangerous`,
which reproduces identically with this patch fully reverted — pre-existing and
unrelated. CI is the authority here, not my laptop.

- [x] Documentation — N/A (internal helper; no user-facing config change)
- [x] `cli-config.yaml.example` — N/A (no new config keys; `docker_extra_args` already exists)
- [x] `CONTRIBUTING.md` / `AGENTS.md` — N/A (no architecture change)
- [x] Cross-platform impact considered — pure string parsing; Windows drive paths
      already handled by the existing `_docker_volume_uses_host_path`, which this reuses
- [x] Tool descriptions/schemas — N/A (no tool behavior change beyond correctly
      prompting where it previously did not)

---

### DIFF
--- a/contributors/emails/dewoller@gmail.com
+++ b/contributors/emails/dewoller@gmail.com
@@ -1,0 +1,1 @@
+dewoller
--- a/tests/tools/test_modal_sandbox_fixes.py
+++ b/tests/tools/test_modal_sandbox_fixes.py
@@ -358,6 +358,52 @@ class TestDockerHostBindApproval:
         assert _tt_mod._docker_has_host_access(
             {"env_type": "modal", "docker_volumes": ["/tmp:/x"]}) is False
 
+    def test_docker_extra_args_host_bind_detected(self):
+        """Host binds expressed as raw docker args also disable the fast path.
+
+        ``docker_extra_args`` is forwarded to ``docker run`` verbatim, so
+        ``-v /host:/c`` mounts exactly what a ``docker_volumes`` entry would.
+        Reading only ``docker_volumes`` lets the identical mount arrive with the
+        approval fast-path still armed, which is the case this guard exists to
+        prevent.
+        """
+        # -v with a host path -> host access.
+        assert _tt_mod._docker_has_host_access(
+            {"env_type": "docker", "docker_volumes": [],
+             "docker_extra_args": ["-v", "/srv/data:/srv/data"]}) is True
+        # --volume long form -> host access.
+        assert _tt_mod._docker_has_host_access(
+            {"env_type": "docker", "docker_volumes": [],
+             "docker_extra_args": ["--volume", "/tmp:/hosttmp"]}) is True
+        # --volume=... joined form -> host access.
+        assert _tt_mod._docker_has_host_access(
+            {"env_type": "docker", "docker_volumes": [],
+             "docker_extra_args": ["--volume=/tmp:/hosttmp"]}) is True
+        # --mount type=bind -> host access.
+        assert _tt_mod._docker_has_host_access(
+            {"env_type": "docker", "docker_volumes": [],
+             "docker_extra_args": ["--mount", "type=bind,source=/etc,target=/etc"]}) is True
+        # Named volume via -v -> not host access.
+        assert _tt_mod._docker_has_host_access(
+            {"env_type": "docker", "docker_volumes": [],
+             "docker_extra_args": ["-v", "myvol:/data"]}) is False
+        # --mount type=volume -> not host access.
+        assert _tt_mod._docker_has_host_access(
+            {"env_type": "docker", "docker_volumes": [],
+             "docker_extra_args": ["--mount", "type=volume,source=myvol,target=/data"]}) is False
+        # Unrelated args -> not host access.
+        assert _tt_mod._docker_has_host_access(
+            {"env_type": "docker", "docker_volumes": [],
+             "docker_extra_args": ["--network", "none"]}) is False
+        # A trailing flag with no value must not IndexError.
+        assert _tt_mod._docker_has_host_access(
+            {"env_type": "docker", "docker_volumes": [],
+             "docker_extra_args": ["-v"]}) is False
+        # Non-list config value must not raise.
+        assert _tt_mod._docker_has_host_access(
+            {"env_type": "docker", "docker_volumes": [],
+             "docker_extra_args": None}) is False
+
     def test_should_skip_container_guards(self):
         """Docker skips only when isolated; other sandboxes always skip."""
         import tools.approval as A
--- a/tools/terminal_tool.py
+++ b/tools/terminal_tool.py
@@ -372,13 +372,61 @@ def _docker_volume_uses_host_path(volume_spec: str) -> bool:
     )
 
 
+def _docker_mount_spec_uses_host_path(spec: Any) -> bool:
+    """Return True when a ``--mount`` spec bind-mounts a host path.
+
+    Docker defaults ``type`` to ``volume``, so an omitted type is NOT a bind.
+    Only ``type=bind`` reaches host state; named volumes and tmpfs do not.
+    """
+    if not isinstance(spec, str):
+        return False
+    fields = {}
+    for part in spec.split(","):
+        if "=" in part:
+            key, _, value = part.partition("=")
+            fields[key.strip().lower()] = value.strip()
+    if fields.get("type", "volume").lower() != "bind":
+        return False
+    return _docker_volume_uses_host_path(fields.get("source") or fields.get("src") or "")
+
+
+def _docker_extra_args_use_host_path(extra_args: Any) -> bool:
+    """Return True when raw ``docker run`` args bind-mount a host path.
+
+    ``docker_extra_args`` is forwarded to ``docker run`` verbatim, so a ``-v``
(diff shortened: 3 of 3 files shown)


alt-glitch:

Related: #77666 covers the overlapping Docker bind-parser surface. This PR adds distinct docker_extra_args handling plus correct default-volume and malformed-input cases; please consolidate the helpers before merge.

tieubao:
Independently hit the same gap while wiring a container backend on v2026.8.31, and landed on the same shape of fix (`_extra_args_have_host_mount` scanning `-v` / `--volume` / `--volume=` / `--mount` / `--mount=` in the docker arm of `_docker_has_host_access`) before finding this PR, so +1 from a second reproduction. Your version is the better one: distinguishing `type=volume` and named volumes (`myvol:/data`) from host binds is the part mine did not do.

One observation for whoever reviews: the same blind spot exists for any other container backend that forwards raw run flags (plugin-registered providers with their own `*_extra_args` key), because only the docker arm consults the structured volumes list. Not a blocker for this PR, just the reason a backend-agnostic hook on the provider might be worth it later.

Test cases I had that yours already covers: inline `--mount=type=bind,...`, `--volume=`, and `--network none` staying `False`. Nothing to add; happy to see this land.

jonameijers:
I hit this same gap independently and opened #109529 before finding this PR — my fault for searching issue titles rather than PR bodies for `docker_extra_args`. Yours is older and, honestly, the better-argued one: grounding it in the maintainer's stated contract from #42750 and in SECURITY.md §2.4 is more persuasive than arguing it from first principles, which is what mine does. You also default `--mount`'s `type` to `volume`, matching docker; I defaulted to `bind` and only corrected it after review.

So rather than carry a competing PR, here are the forms I found that `_docker_extra_args_have_host_bind` doesn't yet catch. All with `docker_volumes: []` and cwd-mounting off, so `_docker_has_host_access` returns `False` and `_should_skip_container_guards` keeps the fast path:

| `docker_extra_args` | current | expected |
|---|---|---|
| `["-v/tmp:/mnt"]` | False | True |
| `["-iv/tmp:/mnt"]` | False | True |
| `["-iv", "/tmp:/mnt"]` | False | True |
| `["--mount", ' "type=bind",source=/tmp,target=/mnt']` | False | True |
| `["--mount", "type=bind,src=.,dst=/mnt"]` | False | True |
| `["--mount", "type=bind,src=..,dst=/mnt"]` | False | True |

Three causes:

**1. pflag attached and combined shorthand.** `arg in ("-v", "--volume")` matches separate tokens only. Docker's flag parser also accepts an attached value (`-v/tmp:/mnt`) and boolean clusters where the value rides the last flag in the cluster (`-iv/tmp:/mnt`, `-iv=/tmp:/mnt`). Scanning for a `v` inside a single-dash cluster and taking the remainder — or the next token when `v` ends it — covers the family.

**2. `--mount` is CSV, not a comma split.** Docker trims the whole value and then parses it as CSV, so a quoted field may contain commas, and a leading space or tab stops the quote landing at position 0 — leaving `"type=bind"` to parse as the literal key `"type`. `next(csv.reader([value.strip()]))` handles both.

**3. Dot-relative sources.** `_docker_volume_uses_host_path`'s prefix list (`/`, `~`, `./`, `../`) misses bare `.` and `..`, which docker resolves against the client working directory. Classifying the source against docker's *named-volume* grammar (`^[a-zA-Z0-9][a-zA-Z0-9_.-]*$`) and treating anything else as a path covers those by construction rather than by extending the prefix list.

I have 55 tests for these in #109529 (`tests/tools/test_docker_host_access_extra_args.py`), including negatives so a sloppier parser doesn't pass, and propagation checks through both `check_dangerous_command` and `check_execute_code_guard`. **They're yours if you want them — happy for you to lift the cases wholesale, and I'll close #109529 once they're folded in here.** One good PR for this beats four partial ones, and triage has already flagged that there are four (#77666 and #78035's internal `9ee310573` are the others).

One design note if you do take them: I ended up making detection deliberately conservative — it flags mount-ish tokens without modelling which flags consume values, so `["--label", "--volume=/tmp:/mnt"]` enables guards on purpose. That trades a documented false positive for closing the family of bypasses. Reviewer feedback on mine was that the tradeoff is acceptable *if disclosed*, since under unattended deny policies a false positive blocks rather than prompts. Your call whether that fits the framing here.

jonameijers:
Following the coordination request on #109529: both PRs address the same `docker_extra_args` approval-introspection gap, so these should be treated as alternatives rather than independently merged fixes. The latest #109529 commit a2c5029962 adds regression cases for `--volumes-from` and local volume-driver bind options, plus real execute_code entry-point assertions under unattended deny. Those tests can be reused here regardless of which parser maintainers choose. #109529 deliberately accepts false positives for opaque volume options and mount-looking flag values; it does not claim precise Docker isolation.

PR #97728, opened 2026-08-29
fix(desktop): let upward wheel escape transcript bottom-follow

alexnick:
## Summary
- Stop bottom-follow immediately when the reader wheels upward, including over nested transcript content.
- Remove the earlier reconcile workaround.
- Add focused regression tests.

## Verification
- 176 thread tests pass.
- TypeScript build passes.
- The latest commit was manually tested on a fresh Windows build; I personally confirmed it fixes the issue.

### DIFF
--- a/apps/desktop/src/app/contrib/hooks/use-background-sync.test.ts
+++ b/apps/desktop/src/app/contrib/hooks/use-background-sync.test.ts
@@ -166,6 +166,25 @@ describe('active transcript refresh', () => {
     vi.mocked(getLatestSessionMessages).mockResolvedValue(transcript('answer') as never)
   })
 
+  it('keeps the mounted transcript when a refresh observes a transient empty page', async () => {
+    vi.mocked(getLatestSessionMessages).mockResolvedValue({ messages: [], session_id: ACTIVE_STORED_ID } as never)
+    const fixture = makeRefresh()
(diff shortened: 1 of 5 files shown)


alexnick:
My bot opened the initial PR before manual testing. Sorry — that was a mistake. The latest commit has now been fully tested by hand on a fresh Windows build, and I personally confirmed it fixes the issue.

alexnick:
Follow-up after manual Windows testing found two refresh races not covered by the earlier scroll-only fix:

- Ignore transient empty persisted-tail reads for both the active transcript and mounted workspace tiles, so an empty poll cannot temporarily clear mounted history.
- Keep a fresh New Session draft local instead of re-fetching model/config defaults after typing has started; profile and gateway transitions still own default reseeding.
- Add focused regression coverage for active transcripts, workspace tiles, and sticky fresh drafts.

Verification:
- `npm run test:ui -- src/app/contrib/hooks/use-background-sync.test.ts src/app/contrib/hooks/use-background-sync.test.tsx` — 29/29 passed
- ESLint on the three changed files — passed
- `npx tsc -p . --noEmit` — passed
- Manually tested on a packaged Windows build: New Session retained focus/caret/draft, and existing Session/Bot Chat refreshes remained stable while reading above the bottom.

PR #97730, opened 2026-08-29
test: replace the source-reading assertions with behavioral ones

### LINKED ISSUE #97729: [Feature]: convert the source-reading tests to behavioral tests
### Problem or Use Case

Five tests across the suite assert on module source text: `inspect.getsource` greps in the compaction tool refresh, the usage anchor invalidation, and the 413 recovery wiring, plus file-substring asserts in the read_file warning test, the anydoc pin lockstep, and the installer helper test.

AGENTS.md bans this shape ("Never read source code in tests"). These tests fail on any innocent reformat of the module and stay green while behavior drifts, because the code they describe is never executed. The 413 wiring test also greps for exact lines inside `conversation_loop`, a file that is edited constantly.

### Proposed Solution

Convert all five to behavioral tests that run the real code:

- compaction refresh contract: drive a real `AIAgent._compress_context` with a stub compressor (harness already proven by `test_compression_logging_session_context.py`), observe the refresh after prompt invalidation, survive a raising refresh, and clear the usage anchor.
- usage anchor invalidation: call the codex native compaction recorder and `AIAgent.reset_session_state` with the anchor set.
- 413 progress: extract the loop's decision into `message_sanitization.compression_recovery_progress` (next to `serialized_messages_bytes`), have the loop call it, and test the function directly, including the 5% boundary and the zero-bytes guard.
- read_file warning and pin lockstep: assert the warning from the real `_needs_ocr_warning` output, parse pyproject with `tomllib`, and import `LAZY_DEPS` instead of regexing both files.
- installer helper: keep the byte-equality drift guard between the two scripts (a real cross-file invariant) and drop the intra-body substring asserts, the live-bash test already executes the helper.

### Alternatives Considered

Leaving them: every future refactor of `conversation_loop`, `conversation_compression`, or the installers risks false-red CI from formatting-only diffs, which is exactly the failure mode AGENTS.md calls out.

Adolanium:
Replace source-text assertions with behavioral coverage for oversized image recovery, compaction tool refresh, installer environment handling, and OCR warnings. The recovery predicate is shared with the current turn_overflow handler. Preserve main's newer usage-anchor tests. Installer tests execute both scripts' sync helpers on Linux.

Fixes #97729

Validation: 32 tests passed via scripts/run_tests.sh; three Linux installer cases are skipped on this Windows host. The five compaction tests passed again after strengthening call-order and failure-path assertions.


Enough1122:

Reviewed `agent/conversation_loop.py`, `agent/message_sanitization.py` + five test files for PR #97730. Replaces five source-reading assertions (`inspect.getsource` greps / file substring checks) with behavioral ones — exactly the shape the project rubric bans ("Never read source code in tests"). This is a strong, directionally-correct change: it extracts `compression_recovery_progress()` as a real function the loop imports, so the loop's decision and the tests cannot drift, and it pins real contracts (usage-anchor invalidation, tool-refresh ordering, pyproject/lazy-deps lockstep via `tomllib`, live-bash helper behavior) instead of formatting.

Non-blocking observations:

1. **Heavier tests now instantiate a full `AIAgent` + real `SessionDB`** (e.g. `test_reset_session_state_clears_the_anchor`, `_commit_path_contract`). This is the intended trade (real E2E over mocks, per the rubric), but these are materially slower and more env-sensitive than the source-grep they replace. With `skip_context_files=True`/`skip_memory=True` and a patched `OPENROUTER_API_KEY` they look hermetic; just watch suite runtime and CI flakiness.

2. **Dead imports in `_commit_path_contract`:** `import tempfile` / `from pathlib import Path` are both marked `# noqa: F401` "kept for harness symmetry". Dead imports add noise for future readers; consider removing them.

3. **`compression_recovery_progress` preserves behavior exactly** (verified: `new_len < original_len or (new_bytes > 0 and new_bytes < original_bytes * 0.95)` matches the original inline expression), so this is a pure refactor with no semantic drift. The boundary test (`exactly 5% is not progress`) is a good pin.

4. **`test_install_sh_uv_lock_config`:** dropping the intra-body substring assertions in favor of the live-bash test is the right call — the remaining assertion that both installers carry an identical helper still guards cross-file drift without testing formatting.

A clean, principled refactor that deletes an entire banned antipattern class. Nothing blocking.

Adolanium:
I updated this on the refactored main and pushed e1c82675da as one commit. I moved the 413 coverage to the current overflow handler and retained main's newer usage-anchor coverage. The compaction test now verifies prompt invalidation, tool refresh, and rebuilding in order, including a refresh failure.

32 tests passed via scripts/run_tests.sh; three Linux installer cases are skipped on this Windows host. The five compaction tests passed again after strengthening call-order and failure-path assertions.

PR #80425, opened 2026-08-06
fix: retain requested branch on no-op update with `--branch`

JonthanaHanh:
## Summary

`hermes update --branch B` unconditionally restores the original branch in the no-op path (commit_count == 0), even when the user explicitly requested branch B. This leaves the checkout on the wrong branch with exit code 0.

## Root cause

In `hermes_cli/update_cmd.py` lines 3862-3869, when `HEAD..origin/{branch}` yields zero commits, the updater:
1. Restores the stash (correct)
2. Checks out the original branch (BUG when `--branch` was explicit)

The non-zero commit path (actual updates) correctly stays on the requested branch — the bug is only in the no-op path.

## Fix

Guard the branch restoration with `getattr(args, "branch", None) is None` — only restore the original branch when `--branch` was NOT explicitly passed. This is the same pattern used by `_cmd_update_check` via its `branch_explicit` parameter.

## Issues fixed

- Fixes #80412 (no-op path restores original branch)
- Fixes #80397 (remote-only branch update discarded)

## Test plan

- [ ] `hermes update --branch B` when B is already current → stays on B, prints "Already up to date!"
- [ ] `hermes update` (no --branch) → behaves as before (stays on main or restores original)
- [ ] `hermes update --branch B` when B has new commits → stays on B, applies updates

PR #97731, opened 2026-08-29
fix(cli): keep `hermes update --branch` on the requested branch when already up to date

kzmx23:
## The Problem

`hermes update --branch NAME` updates against a non-default branch and switches the checkout to it (auto-stashing as needed). However, on the "already up to date" early-return path the command unconditionally switched the working tree *back* to whatever branch the user was on before the update. The net effect: when the requested branch happened to be up to date, `hermes update --branch NAME` silently undid the branch switch the user explicitly asked for, leaving them on their original branch with no indication of what happened.

## The Solution

Record whether `--branch` was passed explicitly and honor that intent on the up-to-date path:

- Skip the "restore the previous branch" checkout when the branch was named explicitly, so the checkout stays on the requested branch.
- Emit a clear confirmation (`Already up to date — now on '<branch>'`) so the resulting branch is never ambiguous.

The change is confined to the up-to-date early-return in the update command; the normal fetch/merge path and the default (no `--branch`) behavior are unchanged. The diff is 9 lines in a single file.

## Why this matters

`--branch` is only useful if the checkout actually lands on the requested branch. Reverting HEAD on the up-to-date path made the flag behave inconsistently depending on whether new commits happened to be present — the same command left the user on different branches for reasons unrelated to their request. Honoring the explicit branch on every path makes `--branch` predictable and its outcome self-evident from the command's output.

### DIFF
--- a/hermes_cli/update_cmd.py
+++ b/hermes_cli/update_cmd.py
@@ -3733,6 +3733,11 @@ def _cmd_update_impl(args, gateway_mode: bool):
         # minutes on a non-single-branch checkout. Fetch only what we update
         # against.
         branch = _m()._resolve_update_branch(args)
+        # An explicit `--branch NAME` is a request to land on that branch, not
+        # merely to fast-forward it. Track that intent so the "already up to
+        # date" early-return below does not switch the checkout back to whatever
+        # branch the user happened to start on.
+        branch_explicit = bool(getattr(args, "branch", None))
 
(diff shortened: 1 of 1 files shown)


alt-glitch:

Duplicate of #80425: both patches preserve an explicitly requested `--branch` on the no-op update path by preventing checkout restoration to the starting branch.

PR #97732, opened 2026-08-29
feat(cli): add `hermes update --merge-ref` to merge one upstream ref

kzmx23:
## The Problem

`hermes update` only ever fast-forwards the local checkout against `origin/<branch>`. When `origin` has no new commits it prints "Already up to date!" and returns early. There is no supported way to pull in one specific ref or tag from the `upstream` remote and merge it into the current branch. Operators running Hermes as a fork therefore cannot land an exact upstream release into their branch through the built-in updater — they have to drop to raw git and reproduce the fetch, tag resolution, and merge by hand, without the updater's rollback safety net.

## The Solution

Add a `--merge-ref <tag-or-ref>` flag to `hermes update`:

- Fetches the single named ref or tag from the fixed `upstream` remote and merges the resolved commit into the currently checked-out branch. The branch stays checked out; the ref is merged into it and HEAD is never detached.
- Runs independently of the `origin/<branch>` fast-forward pull, so it still applies even when `origin` is already current — the "Already up to date" early-return no longer swallows the request.
- Is mutually exclusive with `--check`; argparse rejects both flags at parse time rather than relying on runtime validation.
- Resolves the ref via `FETCH_HEAD^{commit}` so an annotated tag merges the commit it points to, not the tag object.
- Fails closed on conflict: the in-progress merge is aborted, HEAD is reset to its exact pre-update SHA, and the command exits non-zero before any install or dependency step runs. If the restore itself fails, the exact recovery command is printed.

## Why this matters

The updater already understands fork checkouts (it detects a fork and can sync with `upstream`). `--merge-ref` closes the gap for that same audience: a fork operator can bring a chosen upstream ref or release tag into their branch with a single command and keep the updater's fail-closed rollback guarantees, instead of running the fetch-resolve-merge sequence manually and risking a half-applied or conflicted tree.

### DIFF
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -5058,6 +5058,7 @@ from hermes_cli.update_cmd import (  # noqa: F401
     _leftover_pausable_gateway_pids,
     _log_only_write,
     _mark_skip_upstream_prompt,
+    _merge_upstream_ref,
     _npm_bin_exists,
     _npm_lockfile_changed,
     _npm_manifest_paths,
--- a/hermes_cli/subcommands/update.py
+++ b/hermes_cli/subcommands/update.py
@@ -25,7 +25,11 @@ def build_update_parser(subparsers, *, cmd_update: Callable) -> None:
         default=False,
         help="Gateway mode: use file-based IPC for prompts instead of stdin (used internally by /update)",
     )
-    update_parser.add_argument(
+    # --check (read-only status) and --merge-ref (apply-path ref/tag merge)
+    # are mutually exclusive — argparse rejects both at parse time rather than
+    # requiring runtime validation.
+    _check_or_merge_ref = update_parser.add_mutually_exclusive_group()
+    _check_or_merge_ref.add_argument(
         "--check",
         action="store_true",
         default=False,
@@ -59,6 +63,18 @@ def build_update_parser(subparsers, *, cmd_update: Callable) -> None:
             "If the local checkout is on a different branch, hermes will "
             "switch to the requested branch first (auto-stashing any "
             "uncommitted changes)."
+        ),
+    )
+    _check_or_merge_ref.add_argument(
+        "--merge-ref",
+        default=None,
+        metavar="TAG_OR_REF",
+        help=(
+            "Merge this single tag or ref from the 'upstream' remote into the "
+            "current fork branch (fetch upstream <ref>, then merge the "
+            "resolved commit; runs even when origin has no new commits). "
+            "Fails closed on conflict: HEAD is restored to its exact "
+            "pre-update state and no install/dependency step runs."
         ),
     )
     update_parser.add_argument(
--- a/hermes_cli/update_cmd.py
+++ b/hermes_cli/update_cmd.py
@@ -3734,6 +3734,13 @@ def _cmd_update_impl(args, gateway_mode: bool):
         # against.
         branch = _m()._resolve_update_branch(args)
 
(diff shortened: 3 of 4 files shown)


Enough1122:

Reviewed `hermes_cli/subcommands/update.py` + `hermes_cli/update_cmd.py` for PR #97732. Adds `hermes update --merge-ref <ref>` to fetch one tag/ref from the fixed `upstream` remote and merge it into the current fork branch, bypassing the "Already up to date!" early-return (scoped correctly to the `origin/<branch>` comparison). Fail-closed behavior is well-handled: fetch failure, unresolved ref, and merge conflict each restore `pre_pull_sha` and exit non-zero before any install/dependency step runs. `--check`/`--merge-ref` mutual exclusion via an argparse group is the right way to avoid runtime validation.

Non-blocking observations:

1. **No tests added** for the new flag/flow — the diff touches only the two CLI files. The fetch/merge/abort-with-restore logic is exactly the kind of resolution-chain behavior the rubric asks to cover E2E (subprocess-level with a fake git). Consider at least a test for the conflict-abort path restoring `pre_pull_sha`.

2. **Fail-closed restore also discards the origin pull landed earlier in the same run.** If `--merge-ref` conflicts after the `origin/<branch>` ff merge already advanced HEAD, `git reset --hard pre_pull_sha` reverts *both* — the origin update is silently lost and the user must re-pull. That matches the documented "restored to its exact pre-update state" contract, but it's worth a comment or message so a user doesn't expect the origin merge to have survived.

3. **`FETCH_HEAD^{commit}` handling is correct** for annotated tags, and `--no-edit` is a sound non-interactivity assumption (consistent with the rest of the updater). The manual-recovery message prints the right pre-pull SHA on restore failure.

4. **Minor:** `merge_ref = str(getattr(args, "merge_ref", None) or "").strip() or None` — the empty-string→None normalization is defensive and fine; the comment explains the argparse default.

The feature is scoped, documented, and fails closed. Main gap is test coverage for a fairly involved git-flow addition. Nothing blocking.

kzmx23:
Thanks for the thorough pass — really useful.

Addressed the missing-tests note in 62c40ee4a: I extracted the fetch → resolve → merge → abort → restore block out of `_cmd_update_impl` into a small, directly testable helper `_merge_upstream_ref(...)` (behaviour-preserving move — same body, just parameterized project root and `return False` instead of `sys.exit(1)`), and added real-git tests (no fake-git monkeypatching) for the fail-closed contract:
- a conflicting merge aborts the in-progress merge and restores HEAD to the exact pre-update SHA, leaving a clean tree and no `MERGE_HEAD`;
- a clean merge of an **annotated tag** returns True and lands the change, exercising the `FETCH_HEAD^{commit}` peel leg.

On observation #2 (the fail-closed restore also reverts the `origin/<branch>` ff-merge that landed earlier in the same run): that's intentional and matches the documented "restored to its exact pre-update state" contract — a partial state (origin merged, requested ref not) is harder to reason about than a clean revert-and-retry. Good call that it deserves to be explicit; surfacing it in the conflict message is a fair follow-up.

kzmx23:
Follow-up done in ad7803d62: on a `--merge-ref` conflict, the fail-closed restore now prints a note when it also rolled back an `origin/<branch>` update that landed earlier in the same run, pointing the user to re-run `hermes update` to re-apply it. It fires only when HEAD actually advanced this run; the no-origin-advance path stays quiet. Covered by a new test.

PR #97733, opened 2026-08-29
feat(agent): LLM-generated session handoff with insight extraction

kzmx23:
## The Problem

When a long session hits context compaction, or the process restarts, the in-context conversation is discarded and the rolling compression summary is the only thing that survives. That summary is optimized to keep the model coherent mid-turn — it is not a durable, on-disk record a human or a fresh agent can open later to recover *what the session was actually trying to do*: the goal, the decisions taken (and the alternatives rejected), the next steps, which files changed, and what has and has not been verified. There was also no lightweight way to capture the small number of long-lived facts a session produces (user preferences, project facts, "do not do it this way" corrections) so they are not re-learned from scratch every time.

## The Solution

Two small, self-contained modules plus a minimal CLI entry point:

- **`agent/handoff.py`** writes a durable markdown handoff to disk (`handoffs/handoff-<session_id>.md`, with a `latest.md` pointer and a rolling `index.json`). By default the auxiliary "compression" model summarizes the session into fixed sections (Goal / Done / Next steps / Key decisions / Dead ends / Changed files / Current state / Context / Recommended first action). If the LLM call fails, it falls back automatically to a dependency-free *extractive* handoff — a lossless raw transcript excerpt — so a handoff is always produced even during shutdown or an in-progress compression failure. A `strict_llm` mode (used by the manual command) surfaces an LLM failure as `HandoffLLMError` instead of silently writing a lower-fidelity extractive file dressed up as an LLM summary; the caller can then retry or explicitly ask for the extractive record.

- **`agent/insight_extraction.py`** parses and normalizes durable "facts" from an LLM response (validation, clamping, enum coercion, de-duplication). The same single handoff LLM call also emits a machine-parseable durable-facts JSON block, so facts are recovered with **zero extra LLM calls** and routed to an optional `memory_manager` that exposes `ingest_extracted_facts(facts, session_id=...)` (duck-typed — no dependency on any specific memory backend). A standalone extractor is also provided for callers that want facts without generating a full handoff.

- **CLI wiring**: a new `/handoff-session [--extractive] [focus topic]` command triggers a manual handoff from the terminal (strict by default; `--extractive` forces the no-LLM path). This is the only integration point added; the modules are otherwise standalone and importable.

Secrets are never guaranteed safe by this module — the prompt only *asks* the model to omit tokens/passwords; any hard guarantee must come from the consuming memory store's write path.

## Why this matters

Continuity across compaction and restarts is the difference between a fresh agent re-deriving context (slow, lossy, error-prone) and resuming from an explicit map of goals, decisions, and next steps. The extractive fallback makes the handoff dependable exactly when the system is under stress (shutdown, compression failure, auxiliary LLM unavailable), and the strict mode keeps a manual handoff honest instead of quietly downgrading. Folding insight extraction into the same LLM call adds durable-memory capture at no additional inference cost.

Tests: `tests/agent/test_handoff.py` and `tests/agent/test_handoff_strict.py` cover the extractive path, the LLM path, metadata folding, strict-mode failure surfacing, and the read/index helpers.

### DIFF
--- a/agent/handoff.py
+++ b/agent/handoff.py
@@ -1,0 +1,506 @@
+"""Durable session handoff files for compaction and restarts.
+
+Handoffs are an operational continuity layer separate from the in-context
+compression summary: they are written to disk before context is discarded so a
+future continuation can recover goals, decisions, next steps, and changed files.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import re
+import subprocess
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any, Iterable
+
+from hermes_constants import get_hermes_home
+from agent.insight_extraction import (
+    DURABLE_FACTS_PROMPT_SECTION as _DURABLE_FACTS_SECTION,
+    normalize_facts,
+    parse_facts_block,
+)
+
+logger = logging.getLogger(__name__)
+
+HANDOFFS_DIRNAME = "handoffs"
+_DEFAULT_EXCERPT_CHARS_PER_TOKEN = 4
+
+# Sentinel returned by _message_excerpt when a transcript has no usable text.
+# Exposed so callers (e.g. insight_extraction) can detect an empty excerpt by
+# identity instead of matching on a localized substring.
+EMPTY_EXCERPT_SENTINEL = "- No textual context available."
+
+
+def _safe_session_id(session_id: str | None) -> str:
+    raw = str(session_id or "unknown-session")
+    safe = re.sub(r"[^A-Za-z0-9_.-]+", "-", raw).strip("-._")
+    return safe or "unknown-session"
+
+
+def _handoffs_dir(hermes_home: str | Path | None = None) -> Path:
+    root = Path(hermes_home) if hermes_home is not None else get_hermes_home()
+    return root / HANDOFFS_DIRNAME
+
+
+def _stringify_content(content: Any) -> str:
+    if content is None:
+        return ""
+    if isinstance(content, str):
+        return content
+    try:
+        return json.dumps(content, ensure_ascii=False, default=str)
+    except Exception:
+        return str(content)
+
+
+def _message_excerpt(messages: Iterable[dict[str, Any]], *, max_messages: int = 40, max_chars: int = 12_000) -> str:
+    """Return an extractive transcript excerpt suitable for fallback handoffs."""
+    items = list(messages or [])
+    if len(items) > max_messages:
+        head = items[: max_messages // 4]
+        tail = items[-(max_messages - len(head)) :]
+        items = head + [{"role": "system", "content": "... middle messages omitted ..."}] + tail
+
+    lines: list[str] = []
+    used = 0
+    for msg in items:
+        role = msg.get("role", "unknown") if isinstance(msg, dict) else "unknown"
+        content = _stringify_content(msg.get("content") if isinstance(msg, dict) else msg).strip()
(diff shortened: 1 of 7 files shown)


Enough1122:

Reviewed `agent/handoff.py`, `agent/insight_extraction.py`, `cli.py`, `hermes_cli/cli_commands_mixin.py`, `hermes_cli/commands.py` + tests for PR #97733. Adds durable session handoffs (extractive dependency-free default, LLM-summarized opt-in with strict-mode surfacing) plus durable-facts extraction that routes to an optional memory manager with zero extra LLM calls. The design is thoughtful: extractive fallback is genuinely dependency-free (keeps working during shutdown/compression failure), strict mode raises `HandoffLLMError` instead of silently downgrading a manual request, and test coverage is extensive (strict/non-strict, metadata append, index/symlink, excerpt truncation).

Non-blocking observations:

1. **Secrets hygiene in durable facts.** The module docstring explicitly delegates secret-protection to the memory store's write path and merely *asks* the model to omit secrets. `normalize_facts` is the last deterministic filter before long-term storage — a lightweight denylist for common credential patterns (`sk-…`, `ghp_…`, `Bearer …`) there would add defense-in-depth at the choke point. Worth considering given facts land in persistent memory.

2. **`_gather_git_context` runs `git log --oneline -20` / `git status` in `hermes_home`.** If the hermes home happens to be inside a git repo, up to 20 commit subjects get embedded into the handoff prompt and file. If the intent is project context, scoping to the actual project root (or truncating) would be tighter; otherwise worth a comment that home-directory git context is the intended scope.

3. **`_message_excerpt` keeps only head (1/4) + tail for long transcripts** (and max 80 msgs / 25k chars on the LLM path). Key decisions often sit in the omitted middle. An importance/salience-aware sample (or keeping the most tool-dense section) would materially improve handoff quality for long sessions; currently the middle is dropped on both paths.

4. **`_write_latest_pointer` symlink fallback:** on Windows without symlink privilege, `latest.md` becomes a *copy* of the target content and won't track later updates to the same target (stale pointer). Minor, but consider making the fallback re-write on each call (it does) and documenting the staleness.

5. **Cosmetic:** metadata format differs between paths — LLM path appends `platform: telegram` (lowercase, sorted), extractive path uses `Platform: telegram` (capitalized). Also, reusing `task="compression"` for the handoff call means the handoff inherits whatever provider/budget that task type resolves in `auxiliary_client`; `max_tokens=4000` for a full structured handoff may be tight — worth confirming that's intended.

6. **No E2E test** that `memory_manager.ingest_extracted_facts` actually fires on a real parsed facts block (the ingest is unit-tested via mocks). Given the rubric's E2E preference and that this is the one cross-module coupling, a small integration test would de-risk it.

A well-structured, well-tested feature with clean strict/fallback semantics. The points above are hardening suggestions, not blockers.

kzmx23:
Thanks for the detailed read.

1. **Secrets denylist in `normalize_facts`:** agreed it's a sensible defense-in-depth choke point. I'd prefer to keep redaction owned by the memory store's write path as the single source of truth rather than duplicate a credential-pattern list here, so I'll track this on the store side as a follow-up rather than in this PR.
2. **`_gather_git_context` scope:** fair — pulling up to 20 commit subjects from `hermes_home` is broad. Scoping to the actual project root (or truncating) is a reasonable tightening; noting it as a follow-up.
3. **Middle-of-transcript drop:** also fair. A salience-aware / tool-dense sample would materially improve long-session handoffs; out of scope for this PR but a good future improvement.

Appreciate the pass.

PR #97734, opened 2026-08-29
fix(stt,tts): surface plugin load failures instead of silent fallback, and route plugin max_text_length

kzmx23:
## The Problem

Hermes discovers STT and TTS backends as plugins: a provider name in config (`stt.provider` / `tts.provider`) is matched against the registered plugin providers at runtime. When a configured plugin provider fails to import — a broken or half-installed addon — the failure was swallowed, with no hardcoded provider name involved:

- **STT**: the dispatcher caught the exception and returned `None`, and the caller then produced a generic "provider not registered" error — the same message a plain typo yields. A genuinely broken plugin was therefore indistinguishable from an unconfigured or misspelled one.
- **TTS**: the dispatcher's `None` fell through to the Edge default and reported success, so a misconfigured `tts.provider` silently produced Edge audio instead of failing.

Separately, `_resolve_max_text_length` never consulted the registered plugin before returning the generic 4000-character cap, so a TTS plugin declaring a larger limit via the existing `list_models()[].max_text_length` field had its output truncated unless the user duplicated the limit as a config override.

## The Solution

- **Surface plugin load failures instead of swallowing them.** The STT dispatcher returns a named error envelope carrying the original import exception (a broken plugin is now distinguishable from an unconfigured one); the TTS dispatcher re-raises into the standard error path; and a terminal branch in `text_to_speech_tool` rejects a configured provider that is neither a built-in, a command-type entry, nor a registered plugin — instead of silently rerolling to Edge. The built-in and command-provider branches are unchanged and still take precedence, so only genuinely unavailable providers hit the new error.
- **Route the max-text-length capability from the plugin.** Resolution now consults plugin discovery inside `_resolve_max_text_length` and reads the registered provider's own declared limit. No provider name is hardcoded and no new capability field is added — it reuses the existing `list_models()` surface. Priority order is unchanged: user config override, then plugin declaration, then the generic fallback.

## Why this matters

The plugin discovery mechanism is already provider-agnostic; these two fixes make it honest. An operator who installs an STT/TTS addon and points config at it now gets either the plugin they asked for or a clear error naming what went wrong — never a silent substitution — and a plugin's declared limits are respected without copying them into config. No provider-specific code is added, so every current and future backend benefits equally.

### DIFF
--- a/tools/transcription_tools.py
+++ b/tools/transcription_tools.py
@@ -1163,6 +1163,16 @@ def _dispatch_to_plugin_provider(
        plugin as unavailable — **not** ``None`` — because the user
        explicitly opted into this plugin via ``stt.provider`` and the
        generic fallthrough message would be misleading.
+    5. Discovery-failure gating: every call that reaches this function
+       already has an explicitly configured, non-built-in provider name
+       (invariants 1-2 filtered out everything else). So when plugin
+       discovery itself raises (a broken or uninstalled addon that fails
+       to import), that means the plugin the user asked for failed to
+       load. This returns a named error envelope carrying the original
+       exception — **not** ``None`` — so a broken plugin is
+       distinguishable from an unconfigured one and the caller surfaces
+       the real load failure instead of a misleading "not registered"
+       error.
 
     Provider exceptions are caught and converted into the standard
     error envelope (matches the legacy built-in error shapes — the
@@ -1195,9 +1205,19 @@ def _dispatch_to_plugin_provider(
             # recovery pattern.
             _ensure_plugins_discovered(force=True)
             plugin_provider = get_provider(key)
-    except Exception as exc:  # noqa: BLE001 — discovery failure is non-fatal
-        logger.debug("STT plugin dispatch skipped (discovery failed): %s", exc)
-        return None
+    except Exception as exc:  # noqa: BLE001 — surfaced below, not swallowed
+        # An explicitly configured plugin provider that fails to even
+        # load must not be indistinguishable from "no plugin configured"
+        # — return a named envelope instead of a bare None.
+        logger.warning(
+            "STT plugin '%s' failed to load: %s", key, exc, exc_info=True,
+        )
+        return {
+            "success": False,
+            "transcript": "",
+            "error": f"STT plugin '{key}' failed to load: {exc}",
+            "provider": key,
+        }
     if plugin_provider is None:
         return None
 
--- a/tools/tts_tool.py
+++ b/tools/tts_tool.py
@@ -460,6 +460,31 @@ def _resolve_max_text_length(
                 return named_override
             return DEFAULT_COMMAND_TTS_MAX_TEXT_LENGTH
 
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed `tools/transcription_tools.py` + `tools/tts_tool.py` for PR #97734. Surfaces plugin load failures instead of silently swallowing them (STT returns a named error envelope, TTS re-raises so the outer handler reports it), and routes plugin `max_text_length` through the resolver. The behavioral split is internally consistent with each module's contract: STT's dispatcher returns envelopes, TTS's dispatcher raises to the outer `try/except`. The `isinstance(limit, int) and not isinstance(limit, bool) and limit > 0` filtering is correctly defensive.

Non-blocking observations:

1. **The new TTS terminal-error branch changes prior fallthrough behavior.** The `elif provider not in BUILTIN_TTS_PROVIDERS:` branch now returns a hard error where the code previously fell through to the Edge TTS default. The comment asserts the only ways to reach it are "neither built-in, nor tts.providers command, nor registered plugin" — worth confirming the earlier `if`-chain really catches *every* legitimate provider type (custom command providers, any other resolution path), so no valid configuration silently becomes a hard failure. If any provider was previously resolved by a later branch, this would be a regression.

2. **`_resolve_max_text_length` calls `plugin_provider.list_models()`.** If that call raises (broken plugin) the resolver propagates rather than falling back — arguably correct (surface the broken plugin), but the surrounding code has no guard. Confirming the caller handles a raise here (or wrapping in the same "failed to load" treatment) would make the failure path explicit.

3. **STT envelope vs TTS re-raise asymmetry** is documented in each docstring and matches each module's convention — fine, just noting the two tools now fail differently by design.

4. **Minor:** `_ensure_plugins_discovered()` is called non-forced inside the resolver; an installed-but-not-yet-discovered plugin is the documented target case, so this is correct — only noting the timing dependency.

The change fixes a real observability gap (a broken plugin was indistinguishable from "not configured") and routes the TTS cap correctly. The main thing to double-check is observation 1's invariant. Nothing blocking.

kzmx23:
Thanks — careful review.

On the key point (**#1** — does the new `elif provider not in BUILTIN_TTS_PROVIDERS` terminal branch regress any valid provider): I traced the dispatch ladder in `text_to_speech_tool` (`tools/tts_tool.py:2964-3092`). Command providers (`if command_provider_config is not None`), registered plugins (the `_dispatch_to_plugin_provider` walrus `elif`), and every named built-in (elevenlabs / openai / deepinfra / minimax / xai / mistral / gemini / neutts / kittentts / piper) are all handled by branches **above** the new `elif`. It is reached only when the provider is none of those — i.e. explicitly configured but unavailable — which previously fell through to a **silent** Edge reroll. So no valid configuration newly hard-fails; the change only converts that silent fallthrough into a named error, and the final `else` still serves the Edge default for built-ins.

**#2** (`_resolve_max_text_length` propagating a broken-plugin raise): agreed that surfacing the broken plugin is arguably the right behaviour; noting the asymmetry with the guarded paths as a possible follow-up. Thanks!

PR #74743, opened 2026-07-30
feat(junie-acp): JetBrains Junie via ACP — with the self-improvement loop intact (supersedes #69207)

AlexanderPrendota:
## What does this PR do?

Adds **JetBrains Junie** as a native coding-agent provider (`junie-acp`), driven over the **Agent Client Protocol** using the `agent-client-protocol` SDK (already vendored as the `[acp]` extra). Hermes spawns `junie --acp=true` and talks to it as an ACP *client*, exposing Junie through the standard chat surface so it can be selected like any other model/provider.

Junie is an autonomous agent that runs its **own** tools inside the ACP session, and this integration still treats it that way: its native `tool_call` notifications are never fabricated as *pending* OpenAI `tool_calls`, so Hermes never re-runs finished work. What changed since #69207 is that this no longer costs Hermes its own agentic surface:

- Hermes' agent-level tools (the memory/skills self-improvement loop, `todo`) now reach Junie through the ACP **text bridge** — the pattern `copilot-acp` already used, factored out so both clients share it.
- Junie's completed work is **projected into the transcript** as `assistant(tool_calls=[…])` + `tool(result)` rows, so the background-review fork replays real calls, results and errors instead of a one-line activity feed.
- The skill-review nudge counter now moves for iterations that happen inside the provider, and the review fork either works or doesn't spawn.

That addresses @teknium1's review on #69207 in full (details under *Changes Made*). This PR **supersedes #69207**, which was auto-closed under the in-tree-provider policy while the review fixes were being written — see the last section. Rebased on current `main`.

## Related Issue

N/A — new optional provider integration. Supersedes #69207 (closed by hermes-sweeper, not by review). Happy to open a tracking issue if preferred.

## Type of Change

- [x] ✨ New feature (non-breaking change that adds functionality)

## Changes Made

**Provider integration (as in #69207)**

- **New provider client** `agent/junie_acp_client.py`: drives `acp.ClientSideConnection` from a dedicated background asyncio loop, bridged to Hermes' synchronous chat-completions call path. Persistent subprocess reuse and a no-replay-after-prompt-dispatch retry guard (Junie may already have applied side effects).
- **Safety bridge** (client-side ACP callbacks): deny-by-default `session/request_permission` (configurable); sandboxed `fs/read_text_file` + `fs/write_text_file` (path-within-cwd enforcement, write-deny for protected files, secret redaction on read).
[...]

PR #97735, opened 2026-08-29
feat(providers): add ACP subprocess provider to run an ACP agent as a model

kzmx23:
## The Problem

Hermes can talk to REST model endpoints and already ships a Copilot ACP client, but there is no general way to drive a coding agent that speaks the Agent Client Protocol (ACP) over stdio as if it were an ordinary model. Such agents (for example Claude Code via `@agentclientprotocol/claude-agent-acp`) run as a local subprocess and keep a persistent, stateful session across turns, which the request/response REST path does not model.

## The Solution

Add `ACPSubprocessClient` (`agent/acp_subprocess_client.py`): an OpenAI-shim client that spawns an ACP agent as a subprocess and speaks JSON-RPC over stdio. It performs the initialize handshake, creates or resumes a session (`session/new` + `session/load`), sends only the new turn as a delta over what the stateful session already holds, and exposes both a blocking completion and a streaming generator of OpenAI-shaped chunks. Streaming surfaces tool-progress (`session/update` tool_call events) as heartbeat chunks so a consumer's idle watchdog keeps resetting while the agent works silently with tools, and it maps ACP stop reasons to OpenAI finish reasons. The client can resume an existing ACP session via `session/load` when a prior session id is supplied; a failed load falls back to a fresh session and records the failure on usage (`acp_resume_failed`) so a caller can tell the user the live context was lost. Cost/context usage is read from the adapter's own reports.

A model-provider plugin (`plugins/model-providers/claude-agent-acp`) registers a provider profile that routes through this client, mirroring the existing `copilot-acp` plugin. Minimal wiring mirrors the copilot-acp hooks: the runtime client factory dispatches `acp://claude-agent` to `ACPSubprocessClient`, the provider is excluded from the Responses-API upgrade, its blocking completion path is used instead of REST streaming, and it is listed as a provider prefix.

## Why this matters

Reusing the OpenAI-shim contract means the rest of the stack (routing, switching, usage accounting) needs no ACP-specific code beyond the small factory hook, while the stateful subprocess session gives agent-style tools and persistent context that the REST path cannot provide.

## Not included / not verified

- **Session resume is mechanism-only:** the factory forwards `agent._acp_session_id` as `resume_session_id` and the client exposes the live id on `response.usage.acp_session_id`, but nothing in this change populates `agent._acp_session_id` between turns. Without an integrator persisting it, a client rebuild starts a fresh session. No cross-restart persistence either.
- The gateway uses the blocking completion path; a native-streaming render loop with tool-progress display is left as follow-up.
- Unit tests pass (42) and every touched file compiles, but this exact configuration has **not** been exercised against a live ACP adapter.

Tests: `tests/agent/test_acp_subprocess_client.py` (42 passing).

### DIFF
--- a/agent/acp_subprocess_client.py
+++ b/agent/acp_subprocess_client.py
@@ -1,0 +1,959 @@
+"""Generic persistent-session ACP subprocess client (OpenAI-shim).
+
+Unlike ``CopilotACPClient`` — which spawns a fresh ``copilot --acp`` process,
+crams the whole conversation into one prompt, and kills the process after every
+message — this client keeps ONE adapter process and ONE ACP session alive across
+turns of a Hermes session, and sends only the *delta* (the new user turn) each
+call. That matches stateful ACP agents like ``@agentclientprotocol/claude-agent-acp``,
+whose session accumulates context on the adapter side; replaying the full history
+every turn would duplicate context (double token/credit spend).
+
+Design notes
+------------
+* The instance is 1:1 with a Hermes session (it lives on ``agent.client``). The
+  persistent process/session is created lazily on the first ``create()`` and torn
+  down by ``close()`` (session end / idle / provider switch).
+* ``claude-agent-acp`` runs its OWN tools (Read/Write/Bash …) internally; it does
+  NOT delegate file ops to the client via ``fs/*`` and (in ``bypassPermissions``
+  mode) does NOT ask the client for permission. So this shim does NOT inject
+  Hermes tool schemas and does NOT parse ``<tool_call>`` blocks — each ``create()``
+  is one user turn and yields one final assistant text answer.
+* Cost is taken straight from the adapter's ``usage_update`` notification
+  (``cost.amount`` in USD, cumulative per session); we expose the per-turn delta.
+* Permission posture is configurable (default: ``bypassPermissions`` — the
+  operator opted into adapter-side execution with Hermes restricting via cwd/env).
+
+Stage references map to scripts/claude-agent-acp-hermes-plan-v4.md.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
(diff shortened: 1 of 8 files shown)


teknium1:
Thanks for building this. As of #101456 (commit 0cbc6e37ac on main, part 2 of the #74743 split) an ACP / external-process provider no longer needs any core edits: a `ProviderProfile` subclass with `auth_type="external_process"`, the `process_command` / `process_args` / `process_command_env_vars` / `process_args_env_var` fields, and a `create_client()` override is the whole integration. It is discovered from `~/.hermes/plugins/<name>/` (`hermes plugins install owner/repo`) or a pip entry point, passes the registry gate, credential resolution, runtime resolution, the main client builder and the auxiliary client (compression/vision) with zero lines in `hermes_cli/` or `agent/`.

Our policy for vendor integrations is standalone plugins, not entries in the core tree (see AGENTS.md, "Third-party products"), so this PR — which wires the generic ACP subprocess provider into several core files — is superseded by the seam. Closing it; please re-ship the generic ACP subprocess provider as a standalone plugin repo against the seam (docs: `website/docs/developer-guide/model-provider-plugin.md`, section "External-process (ACP) providers"; in-tree example: `plugins/model-providers/copilot-acp/`), and post it in the Nous Discord `#plugins-skills-and-skins`. If you hit anything the seam can't express, open an issue — widening the generic surface is exactly what we want to hear about.

kzmx23:
Thanks for the detailed explanation, and for making external-process providers a first-class seam — that is a much better home for this than core edits. I am already rebuilding the generic ACP subprocess provider against `auth_type="external_process"` / `create_client()` per your guidelines. If I hit anything the seam cannot express, I will open an issue as you suggested.

PR #97736, opened 2026-08-29
feat(agent): master switch to disable the pre-agent auth fallback

kzmx23:
## The Problem

Before the agent runs, the gateway resolves an authentication fallback: if the primary provider's credentials look unavailable, it can silently substitute another configured provider so the turn still goes through. That is convenient, but it is not always wanted. An operator who has deliberately pinned a single provider (for cost, routing, or reproducibility reasons) gets a silent switch to a different model instead of a clear failure, and no supported switch turns that substitution off.

## The Solution

Add a `fallback.enabled` policy reader (`hermes_cli/fallback_config.py`) and honor it at the single pre-agent auth-fallback gate in the gateway:

- When `fallback.enabled` is false, the gateway returns no fallback provider — it does not silently substitute another provider before the agent runs, so the caller gets a deterministic single-provider outcome (and a clear failure if that provider is unavailable) instead of a silent swap.
- The default is unchanged (fallback enabled), so existing behavior is fully preserved. `get_fallback_chain` also gains a defensive guard against a non-dict config.

The change is one gate plus a small config reader; the reader is covered by `tests/hermes_cli/test_fallback_disabled.py` (22 tests).

## Why this matters

Silent provider substitution hides a decision the operator may care about. A master switch makes the behavior explicit and opt-outable without changing anyone's current setup by default.

## Follow-up (not in this PR)

A larger deterministic switch/health-probe pipeline (a transport-aware model health probe and a switch-or-stop policy) exists but is intentionally deferred to a separate PR together with its call sites, to keep this change small and fully wired.

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -61,7 +61,7 @@ from agent.conversation_loop import INTERRUPT_WAITING_FOR_MODEL_PREFIX
 from agent.i18n import t
 from agent.interrupt_compat import request_hard_interrupt
 from hermes_cli.config import cfg_get
-from hermes_cli.fallback_config import get_fallback_chain
+from hermes_cli.fallback_config import get_fallback_chain, fallback_enabled
 
 # --- Agent cache tuning ---------------------------------------------------
 # Bounds the per-session AIAgent cache to prevent unbounded growth in
@@ -2507,6 +2507,11 @@ def _try_resolve_fallback_provider() -> dict | None:
         # root-model normalization now reach the fallback chain too (a raw
         # read here used to miss administrator-pinned fallback_providers).
         cfg = _load_gateway_runtime_config()
+        # Manual mode (master kill-switch): forbid the pre-agent auth fallback —
+        # a primary credential failure must surface, not silently resolve onto a
+        # fallback provider before the agent is even built.
+        if not fallback_enabled(cfg):
+            return None
         fb_list = get_fallback_chain(cfg)
         if not fb_list:
             return None
--- a/hermes_cli/fallback_config.py
+++ b/hermes_cli/fallback_config.py
@@ -2,7 +2,11 @@
 
 from __future__ import annotations
 
+import logging
+
 from typing import Any
+
+logger = logging.getLogger(__name__)
 
 
 def _normalized_base_url(value: Any) -> str:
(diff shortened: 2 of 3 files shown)


Enough1122:

Reviewed `gateway/run.py`, `hermes_cli/fallback_config.py` + tests for PR #97736. Adds a master kill-switch (`fallback.enabled`) with a backward-compatible default of True, wired into the gateway's pre-agent auth fallback (`_try_resolve_fallback_provider` returns None when disabled). The lenient scalar coercion, non-dict guard on the hot construction path, and `get_fallback_chain` non-dict hardening are all defensible; tests exercise the policy readers well.

Non-blocking observations:

1. **Scope vs. docstring mismatch.** `fallback_enabled`'s docstring claims manual mode means "NO automatic route change may occur at any layer — top-level chain, auxiliary fallback, summarizer to main, or pre-agent auth fallback", but this PR only gates the gateway's `_try_resolve_fallback_provider`. If the other layers (auxiliary fallback, summarizer→main, top-level chain consumers) don't consult `fallback_enabled`, then `fallback.enabled: false` won't actually disable them. Either wire those sites in this PR, or scope the docstring/feature claim down to "pre-agent auth fallback" so the docs don't over-promise.

2. **Unrecognized strings fail *open*.** `fallback.enabled: disabled`, `enabled`, `manual`, or any typo not in `_TRUE/_FALSE_TOKENS` silently defaults to the historical enabled=True — the opposite of what a kill-switch's operator intends. Given this is explicitly a safety switch, consider warning on unrecognized values (or treating them as False for a fail-closed posture), rather than silently keeping fallback on.

3. **`resolve_fallback_enabled_default` calls `load_config()` on every implicit agent construction** ("hot construction path"). If `load_config` is not cached, that's per-agent disk I/O on a hot path. Worth confirming it's cheap/cached, or memoizing the policy resolution.

4. **`bool(val)` for `int`/`float`** means `fallback.enabled: 0.0` → False and `2` → True; lenient but reasonable. Consistent with the documented coercion table.

The feature is well-tested and backward-compatible; the main follow-up is closing the gap between the docstring's "all layers" claim and the single site actually wired. Nothing blocking.

kzmx23:
Thanks — both good points.

**#1 (docstring over-promising):** fixed in e20ef647b. Reworded the `fallback_enabled` docstring, the module comment, and the test-module docstring to describe present behaviour — the flag is the policy signal that automatic-routing call-sites consult before changing provider/model, currently wired into the gateway's pre-agent auth fallback (`_try_resolve_fallback_provider`), with the other layers (top-level chain, auxiliary fallback, summarizer-to-main) expected to consult the same flag as each of those call-sites is wired up. Docstring-only; no behaviour change.

**#2 (unrecognized strings fail open):** fair — for a safety switch, silently defaulting a typo'd `fallback.enabled: disbled` back to `enabled=True` is the opposite of operator intent. I kept the current lenient default in this PR to avoid a behaviour change, but a warning on an unrecognized non-empty value (or a fail-closed posture) is a reasonable follow-up — happy to fold it in here if you'd prefer it in-scope.

kzmx23:
Follow-up done in b5bc052c7: an unrecognized non-empty `fallback.enabled` value (a typo like `disbled`, or `manual`) now logs a warning naming the value and the default it kept, instead of silently defaulting. Behaviour is unchanged (still defaults to enabled=True; blank/whitespace stays quiet; recognized values resolve silently) — the point is just to make a misconfigured safety switch visible. Tests added.

Issue #71023, opened 2026-07-24
[Feature]: Live upgrade — update Hermes without killing running subagents (zero-downtime update)


webtecnica:
## Summary

Currently, updating Hermes (via `hermes update`, WebUI auto-update, or cron-triggered update) kills ALL running subagents. This is because subagents are child processes of the parent Hermes — when the parent restarts, children become orphaned zombies.

**The result:** users who click "update" in the WebUI, or have an automatic cron update, lose all in-progress subagent work. Long-running delegations (kanban workers, batch processing, research tasks) are silently killed.

## The Problem

- **WebUI update** → version mismatch → blocks user → restart kills subagents
- **Cron auto-update** → `hermes update` runs → gateway restarts → subagents orphaned  
- **CLI update** → user runs `hermes update` → unaware subagents will die
- **Recovery:** impossible — subagent results lost forever, no checkpoint

### Architecture (blue-green pattern)

```text
BEFORE:     Hermes(old) → [Subagent A, B, C running]
DURING:    Hermes(old) draining → [Subagents finish naturally]
           Hermes(new) ready on staging port
AFTER:     Hermes(new) → socket migration → old exits
```

### Implementation for v1: Checkpoint + Orphan Adoption (~50-100 lines)

1. **Checkpoint file**: before restart, serialize all active subagent state (session_id, goal, pid) to `$HERMES_HOME/state/subagent_checkpoints.json`
2. **Orphan recovery on startup**: new Hermes process scans checkpoint, checks alive PIDs via `os.kill(pid, 0)`, re-attaches via `os.waitpid`
3. **Result collection**: subagents that finish during update — their stdout/stderr are still in the pipe buffer

**Key insight:** orphaned subagents dont die — they keep running until they try to write to a broken pipe. We just pick up their exit status.

## Advantages

| Aspect | Before | After |
|--------|--------|-------|
| WebUI update | Blocks + kills subagents | Zero-downtime |
| Cron auto-update | Jobs lost | Jobs complete naturally |
| Kanban workers | Reset from zero | Continue where they left off |
| Long research tasks | Results lost forever | Results preserved |
| Reliability | Low (update = risk) | High (update = safe) |
| Auto-update adoption | Users hesitate | Fully automatic |

## Related

- #70755 — WebUI auto-reload (complementary)
- #70721 — Plugin update guard (different scope)

---

Would be happy to contribute the implementation if the direction is acceptable.

GottZ:
[...]

PR #57121, opened 2026-07-02
feat(terminal): add local memory guard

oferlaor:
## Summary

- Add an opt-in local terminal memory guard for foreground commands using transient `systemd-run --user` services.
- Wire `terminal.local_memory_max_mb` and tri-state `terminal.local_memory_swap_max_mb` through CLI, gateway, config defaults, and terminal environment parsing.
- Preserve current main's default-on systemd-scope lifecycle for gateway background pipe and PTY workers.
- Let a positive memory setting tighten, never widen, the upstream background bound; add `MemorySwapMax` only when swap policy is explicitly configured.
- Keep child environment values out of argv via owner-only temporary files, with atomic pending→ready launch markers, exactly-once fallback, and cleanup.

Fixes #56865.

## Current-main refresh

- Base: `b6f42c667a59469b477220254fec6e2486e30e51`
- Head: `9c43e91df6158e193be3bb3b54416c5856d4ceb1`
- Current upstream process-registry scope, PTY, kill, checkpoint, and recovery behavior is retained.

## Validation

- `python -m pytest -q -o 'addopts=' tests/tools/test_local_memory_guard.py tests/tools/test_process_registry.py tests/tools/test_terminal_config_env_sync.py tests/tools/test_terminal_compound_background.py tests/tools/test_terminal_tool_pty_fallback.py tests/tools/test_terminal_env_bridge.py` → **136 passed, 4 skipped**
- Ruff on changed files → passed
- `git diff --check origin/main...HEAD` → passed

PR #97739, opened 2026-08-29
feat(gateway): isolate workers and queue under resource pressure

### LINKED ISSUE #70716: [Bug]: Local background terminal executors share the gateway cgroup and can terminate the control plane under memory pressure
## Bug description

When Hermes runs as a long-lived messaging gateway under systemd, local background terminal commands inherit the gateway service cgroup. A memory-heavy executor can therefore cause `systemd-oomd` to kill the entire messaging control plane instead of only the worker command.

This happened in production with a Telegram gateway while a `terminal(background=true, pty=true)` Codex/test run was active. The host remained responsive, but the in-flight turn was silently lost.

## Evidence

Environment:

- Hermes Agent `v0.18.2 (2026.7.7.2)`, upstream `f64e4f4f`
- Linux/systemd, Python 3.11
- 8 GiB RAM, 4 GiB swap
- `hermes-gateway.service`: `MemoryHigh=2500M`, `MemoryMax=3G`, `Restart=always`
- Local terminal backend

At `2026-07-15 21:54:30 +03`, `systemd-oomd` reported:

```text
Path: /user.slice/user-1000.slice/user@1000.service/app.slice/hermes-gateway.service
Pressure: Avg10: 83.30 Avg60: 35.92 Avg300: 9.07
Current Memory Usage: 2.5G
Pgscan: 13904811

Killed .../hermes-gateway.service due to memory pressure for
.../user@1000.service being 72.67% > 50.00% for > 20s with reclaim activity
```

`systemd-cgls --user-unit hermes-gateway.service` showed the gateway Python process and the background shell/Codex/Node/test/MCP descendants in the same cgroup.

Current `main` launches these paths in `tools/process_registry.py::spawn_local()`:

- PTY: `PtyProcess.spawn(...)` around line 727
- pipe mode: `subprocess.Popen(..., start_new_session=True)` around line 781

`start_new_session=True` creates a process session/group, not a resource cgroup.

## Steps to reproduce

1. Run Hermes gateway as a systemd user service with `MemoryHigh`/`MemoryMax` or under a parent monitored by `systemd-oomd`.
2. From a messaging turn, start a memory-heavy local command with `terminal(background=true, pty=true)` (Codex/Claude plus tests is a realistic workload).
3. Verify with `systemd-cgls --user-unit hermes-gateway.service` that the command tree is inside the gateway cgroup.
4. Let the command tree push that cgroup into sustained reclaim/memory pressure.
5. Observe `systemd-oomd` kill the complete gateway cgroup.

## Actual behavior

- The heavy worker and the messaging gateway share one failure domain.
- `systemd-oomd` kills the complete gateway cgroup.
- `Restart=always` can reconnect the platform seconds later, but the active agent turn is not resumed and no failure is delivered to the user.
- From chat, Hermes appears frozen even though the host is alive.

## Expected behavior

- Local heavy executors run outside the gateway control-plane cgroup by default when invoked from gateway mode.
- A worker OOM/limit breach kills only that worker tree.
- The gateway remains responsive and reports the tool failure to the active turn.
- Process cancellation and crash recovery still terminate the tracked worker unit/cgroup without PID-reuse hazards.

## Why this is not a duplicate of #56865

#56865 / PR #57121 adds an **opt-in** memory guard, with `terminal.local_memory_max_mb: 0` disabling it by default. That is useful machinery and likely the right implementation foundation, but after a normal upgrade gateway users remain in the same unsafe shared-cgroup failure domain unless they discover and enable the setting.

The invariant requested here is control-plane isolation for gateway-spawned local background/PTY executors by default. A per-command cap can remain configurable, but the messaging gateway should not inherit worker memory pressure as its own.

## Suggested implementation direction

Build on PR #57121's tracked transient systemd units:

- enable sibling-unit/cgroup isolation automatically for local commands launched from gateway mode on supported Linux/systemd hosts;
- choose/document safe default resource behavior, with config overrides;
- preserve direct-spawn fallback on unsupported systems with an explicit warning;
- cover foreground, background pipe mode, PTY mode, recovered detached sessions, `process kill`, and gateway restart;
- add an integration test for the failure-domain invariant: worker limit breach does not terminate the gateway parent.

## Severity

Suggested: `type/bug`, `comp/gateway`, `tool/terminal`, `backend/local`, `P1`, `sweeper:risk-message-delivery`.

The impact is silent loss of the messaging control plane and active turn, not merely failure of a resource-heavy command.

djpapzin:
Closes #70716
Related: #71023

## Root cause proven on the production VM

This incident was not Linux OOM or cgroup OOM. Kernel logs and the live gateway cgroup `memory.events` contained zero OOM events.

The direct shutdown initiator was the host-local `/usr/local/sbin/hermes-health-guard`. It treated routine SQLite lock contention as fatal and executed `systemctl restart --no-block hermes-gateway.service` after three strikes:

- 2026-08-28 17:25:21 to 17:27:22 SAST: `db_locks` 10, 10, 10, then `HERMES_RECOVERY restart_rc=0`.
- 2026-08-29 08:06:44 to 08:08:45 SAST: `db_locks` 10, 13, 18, then `restart_rc=0`.
- 2026-08-29 11:02:35 SAST: `db_locks=13` reached strike 3/3 while three agents were active. Drain timed out with two agents and four cron jobs; at 11:06:05 systemd exhausted `TimeoutStopSec=3m30s` and SIGKILLed the full `KillMode=control-group` service tree.
- `GatewayRunner.stop()` notified active chats with `Gateway shutting down`, producing the observed Telegram interruption message.

The old guard also restarted the entire gateway for aggregate child memory and raw task-count thresholds. Those are admission and alert signals, not lifecycle authority.

## Design

- Enforce one FIFO, resource-aware admission controller across normal inbound turns, API runs, background-agent turns, and cron jobs.
- Default to three parallel agents, 2048 MiB preserved host/cgroup headroom, and a bounded 16-task queue. Existing running work is preserved; new work receives queue-position/capacity status and starts automatically when a slot becomes available. `/background` tasks now report their queued state to the originating Telegram chat rather than only saying “started.”
- On cgroups-v2 Linux, use the smaller of host `MemAvailable` and gateway-cgroup `MemoryMax - MemoryCurrent` for admission.
- Release only the crashed, cancelled, rejected, or completed task's slot. A synchronous scheduler call on the gateway event-loop thread fails closed instead of deadlocking the control plane.
- Extend independent transient worker scopes to terminal processes, `agent-browser`/Chrome trees, embedded CUA daemons, and stdio CUA MCP processes. Scope environment forwarding preserves names without putting secret values on argv.
- Apply finite `MemoryHigh`, `MemoryMax`, and `MemorySwapMax=0` per worker. `required` mode fails closed if neither a user scope nor the same-UID root-manager scope is available.
- Generate gateway units with `KillMode=mixed` and `OOMPolicy=continue`; feed the systemd watchdog from an independent daemon thread during bounded graceful teardown.
- Persist admission state and structured shutdown evidence, including PID/code/signal/reason, host and cgroup memory, active/queued task IDs, and worker cgroup OOM counters.
- Include admitted `/background` agents in top-level active-work accounting, graceful drain, timeout logs, and cooperative interruption so a maintenance controller cannot mistake a live background agent for an idle gateway.
- Add an alert-only replacement for the host guard. It records pressure, locks, WAL, Telegram connectivity, active work, and queue depth with `recovery_attempted=false`; it never invokes a gateway lifecycle command.
- Add a secret-safe bounded diagnostic command and capped stress command.

The Python `AIAgent` orchestration object remains inside the gateway process in this smallest-change design. Expensive local OS process families are independently capped; uncaught turn exceptions remain isolated by the task boundary. This PR does not claim a new IPC/process-per-turn architecture.

Configuration example:

```yaml
gateway:
  admission:
    max_parallel_agents: 3
    min_host_memory_headroom_mb: 2048
    queue_limit: 16
    poll_interval_seconds: 2
terminal:
  worker_cgroup_mode: required
  local_memory_max_mb: 2048
```

## Exact-head verification

Head: `33e5e7532b49f72538c84f6411b97cc305519ce9`
Base: `a2e19d484cb5591df8dafe667c93345b62d9bf06`
Effective policy: `AGENTS.md`
Effective policy SHA-256: `1a804b27143e4d9b9fc80b4497e6cd05a78a214db2c8195fb4d0504ebaf588f3`

- Canonical exact-tree runner: 382 passed, 6 platform skips, zero failures across 24 affected files.
- Ruff and `git diff --check`: clean.
- Controlled stress: six workers, limit three, peak active three, three queue notices, one deliberate worker crash, queued work resumed and drained, zero unexpected failures, parent/control-plane test process survived.
- Regression coverage includes shared API/cron/background admission, Telegram background queue status, background active-work/drain/interrupt accounting, FIFO resume, queue limits, memory-headroom admission, worker crash/cancel release, duplicate ownership, gateway-loop deadlock prevention, independent worker scope environment, browser/CUA scope wrapping, diagnostics, and alert-only guard decisions.
- Live read-only diagnostics: gateway PID 4179876, RSS 491592 KiB, service cgroup current 4994981888 bytes, high 12884901888, max 17179869184, and `oom=0`, `oom_kill=0`.
- Live alert-only dry run: `decision=healthy`, `recovery_attempted=false`, Telegram connected, active agents zero; gateway PID remained 4179876 and `NRestarts=0`.
- VM user-scope probe: unavailable for `hermes-runtime` because no user bus. Same-UID system-scope probe passed, including environment preservation without values on argv.
- Exact-head disposable OOM proof: scope `hermes-worker-boundary-proof-107860-d09b31c8.scope` at 16 MiB was SIGKILLed by `CONSTRAINT_MEMCG`; kernel evidence names only that worker cgroup. Gateway PID remained 4179876, `NRestarts=0`, and gateway-cgroup `oom=0`, `oom_kill=0`.

## Production boundary

No live source, gateway unit, gateway config, health-guard executable/drop-in, Telegram state, or production process was modified. The new guard was only executed with a `/tmp` state file; it did not control the service. Bounded stress created only unique disposable worker scopes; one initial throttled proof scope was identified and stopped explicitly, and the corrected exact-head proof self-collected after its expected cgroup OOM.

Deployment remains blocked until this exact head receives the required review/check gates. GitHub currently reports no reviews and no status-check rollup for `33e5e753`; the prior head's CI, Docker Build, and Nix flake workflows all required maintainer approval and ran zero jobs, so those stale results cannot gate this rebased head. A maintainer must approve/retrigger the fork workflows and review the exact head. At deployment time, unrelated workloads must be genuinely drained; the guard replacement and gateway restart must use the controlled runbook. Final acceptance still requires live Telegram proof of concurrent running/queue/resume/status, isolated worker failure, PID stability, memory/cgroup bounds, and no restart loop.

## Rollback

1. Stop new admission and wait for active and queued work to drain; do not cancel healthy unrelated work.
2. Restore the timestamped health-guard executable and remove only the documented alert-only drop-in.
3. Revert the deployed code/config commit.
4. Perform one controlled restart under the deployment lock.
5. Verify Telegram response, gateway PID stability, service result, cgroup events, and no restart loop.

Do not remove worker memory limits or raise concurrency as an emergency workaround. Lower `max_parallel_agents` if pressure remains.


djpapzin:
Resumed safety-gate audit (2026-08-29 11:01 SAST): rebased cleanly onto current main. Exact head is a4c449451902e83abdc68ae7863a559ea181d6de; base is aff5125f8edf5095aef5d3d79bbbb101c95b9413. Production deployment remains intentionally blocked: the unchanged live gateway reports 3 active agents, ~12.9 GB cgroup memory, ~1000 tasks, and repeated SQLite-lock health-guard strikes. Gateway cgroup memory.events remains oom=0 / oom_kill=0 and NRestarts=0. No live configuration, service, or workload was modified; fresh exact-head tests/review remain pending until VM pressure is safe.

djpapzin:
Fresh production reproduction (read-only audit, no operator restart): at 11:02:35 SAST the unchanged health guard reached db_locks=13, strike 3/3 and restarted the gateway while runtime active_agents=3. The gateway entered draining, later reported 2 active agents plus 4 in-flight cron jobs, and marked cron IDs 45b7d8caee4b, 65269c78385b, 881ebea630ea, and 7ea9aa00930d interrupted. At 11:06:05 the 3m30s stop timeout expired and systemd sent SIGKILL to the full KillMode=control-group tree. Cgroup memory.events before the restart remained oom=0, oom_kill=0. The old build restarted as PID 4179876 and reconciled the four cron executions as unknown. This is a fresh exact reproduction of the reported interruption path and confirms the cause is health-guard/systemd lifecycle coupling, not OOM. No deployment was attempted. Exact head a4c449451902e83abdc68ae7863a559ea181d6de validation after rebase: canonical runner 237 passed, 6 platform skips, with one unrelated existing Slack config test failing because optional aiohttp is absent; admission-only selection passed; Ruff clean; capped stress 6/3 peaked at 3 with 3 queued notices and gateway parent survival.

djpapzin:
Exact-head update: `34f5050d08ad39a7b630756fe2d8118f4c8e969e` (base `aff5125f8edf5095aef5d3d79bbbb101c95b9413`; effective `AGENTS.md` hash `1a804b27143e4d9b9fc80b4497e6cd05a78a214db2c8195fb4d0504ebaf588f3`).

Additional fixes on this head:

- one global admission boundary now covers normal turns, API runs, background turns, and cron jobs;
- expensive browser/Chrome and CUA process trees now join independently bounded worker scopes;
- bounded defaults are 3 parallel / 2048 MiB headroom / 16 queued;
- diagnostics report per-worker cgroup memory and OOM events;
- the repository includes the documented alert-only replacement for the restart-on-lock host guard.

Exact-tree verification: 289 passed, 6 platform skips, 0 failed; Ruff and `git diff --check` clean. Controlled 6-worker stress peaked at 3 active, emitted 3 queue notices, had zero worker failures, and kept the parent/gateway test process alive.

Read-only VM proof after the tests: live gateway PID `4179876`, `NRestarts=0`, service cgroup `oom=0`, `oom_kill=0`. The new guard was dry-run only with `/tmp` state and recorded `recovery_attempted=false`; no live code/config/unit/process was changed.

Maintainer exact-head review and required gates are requested. The PR remains `mergeStateStatus=BLOCKED` with no reviews. CI, Docker Build, and Nix flake were dispatched for this exact head but each concluded `action_required` with zero jobs, so maintainer workflow approval is also required. The PR has not been merged or deployed. Live Telegram concurrency/queue/failure acceptance remains a post-merge controlled-window gate.

djpapzin:
Exact-head evidence refresh after rebasing onto current `origin/main`:

- Head: `33e5e7532b49f72538c84f6411b97cc305519ce9`
- Base: `a2e19d484cb5591df8dafe667c93345b62d9bf06`
- Effective policy: `AGENTS.md`, SHA-256 `1a804b27143e4d9b9fc80b4497e6cd05a78a214db2c8195fb4d0504ebaf588f3`
- Canonical affected suite: 382 passed, 6 platform skips, 0 failed across 24 files; Ruff and `git diff --check` clean.

Audit follow-ups on this head:

- `/background` now sends an explicit queued-capacity/position notice to its originating chat and resumes automatically.
- Admitted background agents are included in top-level active work, graceful shutdown drain, timeout diagnostics, and cooperative interruption.
- Controlled six-worker stress held peak active to 3, queued 3, deliberately crashed one worker, resumed/drained the queue, and preserved the control-plane process.
- New bounded cgroup proof placed a 64 MiB allocation in a disposable 16 MiB worker scope. The kernel recorded `CONSTRAINT_MEMCG` against `hermes-worker-boundary-proof-107860-d09b31c8.scope`; only that child was SIGKILLed. Live gateway PID stayed `4179876`, `NRestarts=0`, and its cgroup remained `oom=0`, `oom_kill=0`.

No gateway source/config/unit/guard was deployed or restarted. All review and workflow approval evidence from the prior head is stale; maintainer approval/review is required again for this exact head before merge or production acceptance.

djpapzin:
@codex review

Please review exact current head `33e5e7532b49f72538c84f6411b97cc305519ce9`.

djpapzin:
@codex review

Please review exact current head `2ce15b344010ab37ba706194ddc2d3419a3c6025` after rebase onto `main` `9d9f44d63826b18503f44c754e48e1f4f83b3a6e`.

Fresh exact-head evidence: 279 passed, 6 platform skips, 0 failed across all 13 changed test files with the pinned Python 3.11 environment and Slack extra; Ruff and `git diff --check` clean. Controlled stress: 6 workers, limit 3, peak active 3, 3 queue notices, one deliberate worker crash isolated, queue resumed/drained, and control-plane process survived. No production process or service was modified.

andrexibiza (commented):
Reviewed exact head `2ce15b344010ab37ba706194ddc2d3419a3c6025` against `main@9d9f44d63826b18503f44c754e48e1f4f83b3a6e`; the PR base is the current main object at review time. I read the resource-admission/cgroup/shutdown changes, the changed test surface, the existing evidence comments, #70716/#71023, and the live adjacent work in #57121, #97722/#97723, and #97825. There were no existing submitted reviews on this head, so this is not repeating an unchanged-head review.

There is a lot worth preserving here. The central `AgentAdmissionController`, FIFO queue, bounded queue notices, API/cron/background integration, `/background` drain/interrupt accounting, cgroup-v2 host-vs-service headroom witness, worker `MemoryHigh`/`MemoryMax`/`MemorySwapMax=0`, same-UID system-manager scope support, alert-only health guard, and structured shutdown evidence are all the right class of mechanisms. The PR also correctly separates the proved production restart initiator (the host health guard) from OOM claims instead of laundering correlation into causation. The exact-head author receipt of 279 passing tests plus the controlled 6-worker/3-slot crash-recovery stress is useful evidence.

I do see two P1 correctness/ownership blockers before this can close #70716.

### P1 — the default `auto` mode still falls back into the unsafe gateway cgroup on the exact production topology this PR measured

`hermes_cli/config_defaults.py` makes `terminal.worker_cgroup_mode` default to `"auto"`. In `tools/process_registry.py`, however, `_worker_scope_backend()` currently does:

```python
if mode in {"auto", "user", "required"} and _systemd_run_user_scope_available():
    return "user"
if mode in {"system", "required"} and _systemd_run_system_scope_available():
    return "system"
return None
```

So `auto` participates in the user-scope probe but **not** the system-scope fallback. That matters because the PR's own exact-host evidence says the `hermes-runtime` user scope is unavailable because there is no user bus, while the same-UID system scope passed.

On that host, the default configuration therefore resolves `auto -> None`. The background/PTTY `spawn_local()` paths then warn and spawn directly when mode is not `required`, leaving the worker in the gateway service cgroup. The foreground `LocalEnvironment._run_bash()` path likewise gets `backend is None` and does not wrap. In other words, a normal upgrade with the new defaults can still reproduce the core #70716 failure domain even though the PR has already proved that the system-scope escape hatch works on that machine.

That is exactly the distinction #70716 calls out: an opt-in guard is not closure if users remain unsafe after an ordinary upgrade unless they discover/configure the setting.

**Required fix:** make the default policy actually use the available safe backend — e.g. `auto` should try user scope first and then the same-UID system scope — or choose an equivalent default that guarantees isolation when the supported system backend is available. Please add the missing regression for `mode=auto`, user scope unavailable, system scope available => system scope selected, and exercise the foreground plus PTY/pipe spawn surfaces through that decision. `required` failing closed is good, but it does not make `auto` safe.

Until that is repaired, `Closes #70716` is too strong.

### P1 / merge-order — `/handoff` is still a second full-agent concurrency owner outside the new global admission controller

The new controller admits normal inbound turns, API runs, `/background`, and cron. It does not own the handoff watcher. The live #97722 / #97723 pair independently proves the other side of this shape: `_handoff_watcher` fire-and-forgets one **full agent turn plus delivery** per pending row; five pending rows reach five concurrent turns on main. #97723 currently fixes that with a separate closure-local semaphore of 3.

Those two mechanisms should not land independently as permanent peers. If #97739 is the repository-wide resource-admission owner, landing #97723 unchanged would create two unrelated budgets: up to 3 admitted normal/API/background/cron turns **plus** up to 3 handoff turns. That recreates the capacity-boundary split this PR is trying to eliminate, and `admission.active/queued` would under-report real full-agent work.

**Required fix / merge order:** route handoff dispatch through this same `AgentAdmissionController` (or prove that the handoff turn enters another already-admitted surface before doing work). Preserve #97723 / Adolanium's reproduced bug and regression credit, but fold the handoff capacity decision into the one global owner rather than keeping `_HANDOFF_MAX_CONCURRENT` as a parallel authority. #97723 is complementary defect evidence, not duplicate work to discard.

### Attribution / adjacent ownership

There is also live predecessor overlap that should be explicit before merge. #57121 (`oferlaor`) already owns an opt-in foreground local-memory guard and the `terminal.local_memory_max_mb` configuration surface. This PR is broader and can supersede the architectural need for parts of that implementation, but it should do so **with credit and a clear merge-order decision**, not silently duplicate the same config/foreground-guard boundary. The important distinction is: #57121 is a narrower credited precursor; #97739 is the broader control-plane/resource-admission design.

#97825 is different: it is complementary CPU-weight isolation for dispatched kanban workers using delegated child cgroups under the gateway unit. It is not a substitute for these transient terminal/browser/CUA memory scopes. The cgroup topology and cleanup semantics do need to compose, especially if both mechanisms are active under the same service, but contributor ownership should remain separate. #71023 is likewise adjacent live-upgrade/drain work, not closed by this PR.

### Verification gate

The exact current head has three hosted workflow runs, all still `action_required` with zero jobs executed:

- CI `33250909016`
- Nix flake check `33250908682`
- Docker Build, Test, and Publish `33250908686`

So the 279-pass local/exact-tree receipt and VM stress evidence are valuable, but they are not a hosted exact-head green receipt yet. After the two P1 repairs/rebase, the final object still needs fresh CI/Docker/Nix execution on that exact SHA.

The core work here is strong and materially improves the gateway's failure containment. In particular, proving the health-guard restart source, adding a real admission owner, and testing an actual worker memcg kill without killing the gateway are the kind of evidence this class needs. Fix the default backend selection and absorb the handoff concurrency boundary into the same owner, preserve the #57121/#97723 contributor lineage, then rerun the exact-object gates. That would make this a much cleaner closure of the resource-pressure class. 🚀

PR #97741, opened 2026-08-29
fix(agent): retry pre-delivery ReadError UNEXPECTED_EOF with fresh pool

### LINKED ISSUE #97184: Streaming API calls intermittently fail with RemoteProtocolError/SSL EOF across multiple models — stale pooled connections behind Cloudflare
In `agent/relay_llm.py` and `agent/chat_completion_helpers.py`, streaming API calls to OpenRouter intermittently fail with `httpx.RemoteProtocolError: Server disconnected without sending a response` or `httpx.ReadError: [SSL: UNEXPECTED_EOF_WHILE_READING]`, even immediately at the start of a fresh session (as early as message 2), across multiple different models and OpenRouter providers (confirmed on `moonshotai/kimi-k3`, `anthropic/claude-haiku-4.5`, and `anthropic/claude-sonnet-5` — ruling out any single model or backend as the cause).

**Root cause:** OpenRouter sits behind Cloudflare, which enforces its own idle-connection timeout server-side. The `httpx` client (via its underlying `httpcore` connection pool) keeps HTTP/1.1 keep-alive connections open for reuse between requests. If a pooled connection has been idle longer than Cloudflare's server-side timeout, Cloudflare silently closes it without notifying the client. The next request that reuses that stale pooled connection gets either a bare TCP EOF or an empty response — the client has no way to know the connection was already dead until it tries to use it.

This surfaces as a hard failure after 3 retries with short backoff (2-9s), even though a fresh (non-pooled) connection on a subsequent manual retry typically succeeds, since the retry sometimes lands on a new pooled connection.

**Suggested fix:** configure the `httpx.Client`/`AsyncClient` instance used in `relay_llm.py`'s `stream_factory`/`_open_stream` with explicit connection pool limits that recycle connections before Cloudflare's idle timeout:

```python
import httpx

limits = httpx.Limits(
    max_keepalive_connections=5,
    max_connections=10,
    keepalive_expiry=30.0  # seconds — lower than Cloudflare's server-side idle timeout
)
client = httpx.Client(limits=limits, ...)
```

Additionally, the retry logic in `chat_completion_helpers.py`'s `_call` should distinguish this failure mode (`RemoteProtocolError` / `SSL: UNEXPECTED_EOF_WHILE_READING` on connection reuse) and force a fresh, non-pooled connection on retry rather than potentially reusing another stale connection from the same pool. Consider also enabling httpx's transport-level `retries` parameter as a lower-level safety net alongside the existing application-level retry loop.

**Reproduction:** Ran long agentic sessions (tool-heavy, 30+ minutes) and short fresh sessions (2 messages) through Hermes CLI v0.20.4-0.20.6, OpenRouter provider, on a wired ethernet connection with no VPN/proxy in the path (confirmed via `ip a`/`ip route`, no Tailscale active). Same exact failure signature recurs regardless of model or session length. Logs from `~/.hermes/logs/errors.log` available on request.

vsd2807:
## What does this PR do?

Fixes a streaming retry gap where `httpx.ReadError("SSL: UNEXPECTED_EOF_WHILE_READING")` from a stale pooled connection was not classified as a transient connection error.

`RemoteProtocolError` was already handled by the retry path, but `ReadError` was missed, causing the pre-delivery stream to fail without retrying on a fresh connection.

## Related Issue

Fixes #97184

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- Added `httpx.ReadError` to the existing connection-error classification in `agent/chat_completion_helpers.py`.
- Added `unexpected_eof` to the existing SSE error matching for the wrapped `APIError` case.
- Added regression tests covering the `ReadError` retry, the wrapped error case, existing `RemoteProtocolError` behavior, and non-transient errors.
- Added the required contributor attribution for the commit email.

The existing retry path is used, so `keepalive_expiry` remains at 20 seconds and there are no changes to transport retries or connection-pool configuration.

## How to Test

The failure was reproduced locally on the unmodified code by simulating the httpx.ReadError produced by a stale connection:

httpx.ReadError("SSL: UNEXPECTED_EOF_WHILE_READING")

during the pre-delivery streaming path.

Before the fix, the error was not classified as a connection error and the stream failed without retrying.

With the fix, the error enters the existing retry path, the affected client is cleaned up, a fresh client is created, and the retry succeeds.

Regression tests:

pytest tests/run_agent/test_stale_connection_retry_97184.py -v
4 passed

Additional relevant tests:

pytest tests/run_agent/test_create_openai_client_reuse.py -v
6 passed

pytest tests/run_agent/test_stream_stale_circuit_breaker.py -v
4 passed

pytest tests/run_agent/test_request_client_reuse_abort_races.py -v
6 passed

ruff check agent/chat_completion_helpers.py tests/run_agent/test_stale_connection_retry_97184.py
All checks passed

The full `tests/run_agent/test_streaming.py` run had 31 passing tests and 4 skipped, with 3 Anthropic-provider tests unable to run because the optional `anthropic` dependency was not installed in the local environment. Those failures occur during provider initialization and are unrelated to this change.


Enough1122:

Reviewed `agent/chat_completion_helpers.py`, contributor email file + `tests/run_agent/test_stale_connection_retry_97184.py` for PR #97741. Fixes a real retry gap: `httpx.ReadError("SSL: UNEXPECTED_EOF_WHILE_READING")` from a stale pooled connection was not classified as a transient connection error, so pre-delivery streams failed hard instead of retrying on a fresh pool. The fix adds `_httpx.ReadError` to `_is_conn_err` and `"unexpected_eof"` to both SSE phrase lists. Tests are thorough: baseline RemoteProtocolError retry, ReadError UNEXPECTED_EOF retry with a proven fresh client, wrapped `APIError` retry, and a non-transient no-retry sanity check.

Non-blocking observations:

1. **`_httpx.ReadError` is added wholesale, not just the EOF variant.** This classifies *all* read-level failures as transient, including ones that aren't stale-connection EOF (e.g., a malformed mid-stream response). Retries are bounded by `HERMES_STREAM_RETRIES` and use fresh pools, so risk is low — but the classification is coarser than the specific bug. If a distinct non-EOF ReadError family exists that shouldn't retry, the tuple could be tightened.

2. **Fresh-client proof via `id(m)`** (`created_ids[0] != created_ids[1]`) is slightly fragile (id reuse) but acceptable in tests; the call-count assertions are the load-bearing checks.

3. **`_make_agent` instantiates a real `AIAgent`** per test with `skip_context_files`/`skip_memory` — consistent with the project's E2E preference; fine.

4. **Minor:** the contributor email file (`contributors/emails/vaibhavdahiya28@gmail.com`) accompanies the fix — matches repo convention.

The fix addresses the reported failure mode precisely and is well-covered. Nothing blocking.

kshitijk4poor:
Thanks @vsd2807 — current `main` already fixes this via 2355d593f3 ("fix(agent): stale-killed stream unwedges its reader and reconnects (#110769)"), so this PR is superseded and I'm closing it. Your PR predates that commit, so credit to you for pinning it first.

PR #97743, opened 2026-08-29
fix(weixin): preserve inbound preprocessing order per session

xmhuangzhijun-hue:
## What does this PR do?

Weixin's poll loop preserves the order returned by `getUpdates`, but immediately turns every message into an independent task:

```python
for message in response.get("msgs") or []:
    asyncio.create_task(self._process_message_safe(message))
```

For media messages, `_process_message()` then awaits image/video/voice download **before** calling `BasePlatformAdapter.handle_message()`. The generic base adapter correctly serializes agent turns by `session_key`, but it cannot preserve an order it never saw: if image A downloads slowly and the later image B downloads quickly, B reaches the session guard first.

This adds a small per-session preprocessing gate in the Weixin adapter. Its key is produced by the same `build_session_key()` inputs as the generic inbound guard, so it preserves ordering at the correct conversation boundary instead of serializing every chat from one sender. Different sessions still preprocess concurrently.

The registry is reference-counted and removes idle entries. A plain `if not lock.locked(): pop()` is unsafe because `asyncio.Lock.release()` wakes a waiter before that waiter gets CPU; the releasing task could otherwise remove a lock that already has a queued user and let a third message create a second lock.

This only orders the adapter preprocessing stage through entry into `handle_message()`. It does not wait for the agent reply; the existing generic session guard continues to own turn queuing, interruption, command bypasses, and delivery order.

## Related Issue

I could not find an existing Weixin/iLink issue or PR for inbound preprocessing order after searching open, closed, and merged records for ordering, concurrency, sender/session locks, media races, and `getUpdates` task ordering.

## Type of Change

- [ ] ✨ New feature (non-breaking change adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests
- [ ] ♻️ Refactor

## Changes Made

- `gateway/platforms/weixin.py`
  - derive the preprocessing gate from the generic gateway session key
  - preserve same-session poll order before media download can reorder it
  - reference-count and remove idle locks
  - retain concurrency across different sessions
- `tests/gateway/test_weixin.py`
  - same-session slow-first preprocessing remains ordered
  - the same sender in two different rooms still runs concurrently
  - a failed first message releases the queue for the next message
  - every case asserts the lock registry returns to empty

## Testing

The new tests were run against current `main` before the source change: **3 failed**. With the fix: **3 passed**.

```text
scripts/run_tests.sh tests/gateway/test_weixin.py -k InboundPreprocessingOrder
3 passed, 0 failed
```

Ruff passes on both changed files. The broader three-file Weixin run completed **38 passed / 7 failed**; all seven failures are existing async tests that this local shared venv cannot collect because it lacks `pytest-asyncio` (the focused tests use `asyncio.run` and execute normally).

### DIFF
--- a/gateway/platforms/weixin.py
+++ b/gateway/platforms/weixin.py
@@ -1192,6 +1192,12 @@ class WeixinAdapter(BasePlatformAdapter):
         self._send_session: Optional[aiohttp.ClientSession] = None
         self._poll_task: Optional[asyncio.Task] = None
         self._dedup = MessageDeduplicator(ttl_seconds=MESSAGE_DEDUP_TTL_SECONDS)
+        # getUpdates returns an ordered batch, but each message is processed in
+        # its own task.  Media download happens before BasePlatformAdapter's
+        # session guard, so a faster later download can otherwise enter that
+        # guard first.  Refcounts let idle entries disappear without racing a
+        # waiter that has already selected the same lock.
+        self._inbound_order_locks: Dict[str, Tuple[asyncio.Lock, int]] = {}
 
         self._account_id = str(extra.get("account_id") or _wx_secret("WEIXIN_ACCOUNT_ID", "")).strip()
         self._token = str(config.token or extra.get("token") or _wx_secret("WEIXIN_TOKEN", "")).strip()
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed `gateway/platforms/weixin.py` + tests for PR #97743. Preserves inbound preprocessing order per session: the poll loop returns an ordered batch, but each message became an independent task whose media download (before the base adapter's session guard) could reorder. The fix adds a per-session-key refcounted `asyncio.Lock` dict in `_process_message_safe`, serializing preprocessing within a session while keeping different sessions concurrent. Tests cover same-session ordering, cross-session concurrency, and failure-release.

Non-blocking observations:

1. **Per-session serialization also serializes media downloads.** Within one session, downloads that previously overlapped now run one-at-a-time (that's the point — order preservation), but a media-heavy chatty session gets serialized download latency. Correctness-over-throughput is the right call here; just noting the throughput cost is now explicit and per-session.

2. **The refcount pattern is more complex than a plain lock dict, and correctness depends on it.** The increment-before-`try`, decrement-in-`finally`, and `current[0] is lock` guard are all correct (a task holding a refcount prevents pop; the identity check defends against replacement). A short comment on the *waiter-race* (why refcounts rather than `dict.setdefault`) — the existing comment touches it but the invariant is subtle enough to warrant a few more words or a tiny helper, so future edits can't accidentally break it.

3. **`_inbound_order_key` recomputes the session key per message** by re-running `build_source` + `build_session_key` with the same params the generic guard uses. Good — it guarantees the two keys stay consistent by construction (both use the same shared helpers), so the lock scope exactly matches the base guard's session scope. No drift risk identified.

4. **Minor:** `test_failed_preprocessing_releases_session_queue` raises a `RuntimeError` from the fake, and the refcount cleanup in `finally` runs on both the exception and success paths — covered.

The fix is correct and well-tested for a real reordering bug. Nothing blocking.

PR #97744, opened 2026-08-29
feat(bot-mode): add same-gateway Group Chat runner

dokterdok:
## The user problem

Layer 1 gives a Group Chat one durable identity and ordered history, but it does
not run Bots. Today the Desktop renderer still decides which member works next.
If that renderer closes, a Group Chat with every Bot already on one always-on
gateway still cannot continue by itself.

This PR adds the same-gateway execution layer: one gateway owns the bounded
discussion loop, durable member tasks, recovery, and Stop behavior for Bots
that already live on that gateway.

> [!IMPORTANT]
> This PR is stacked on #97712. Review only the commits after the current
> #97712 head `f5908c18d8`. It adds the runner behind the existing `groups.*` contract; the
> normal Desktop creation UX is a separate layer.

## What this layer provides

- One process-owned Group Chat service supervised for the gateway lifetime.
- One hidden `Group: <room-id>` session per member profile, using that Bot's
  existing model, credentials, tools, and workspace.
- Durable task identity, execution generation, cancellation generation, and
  leased single-owner execution.
- Idempotent admission, settlement, failure, cancellation, restart recovery,
  and stale-result rejection.
- A bounded discussion policy: a human message has responders, while later
  Bot-to-Bot work requires an explicit named handoff.
- Offline or uncertain members can be deferred without blocking unrelated
  healthy participants; Retry is explicit.
- Durable Stop intent fences later work before cancellation begins.
- Bounded policy checkpoints, terminal history, per-room scheduling, and
  recovery attempts prevent one large or stalled Group Chat from monopolizing
  the gateway.
- Idle discovery uses a read-only five-second fallback plus immediate wakeups;
  active turns use a short poll and a configured deadline, so four stuck rooms
  cannot wedge every worker slot indefinitely.
- Mutating `groups.*` calls fail closed when the worker is unavailable instead
  of accepting work that nobody will run.

## Deliberate boundaries

- Same-gateway members only.
- No cross-gateway peer/API transport.
- No Desktop continuity client or creation-flow changes.
- No files, Bot-generated artifacts, or messaging-adapter commands.
- No automatic authority failover.

Those capabilities remain separate layers in #97681. This PR can be reviewed
or cherry-picked without accepting them.

## Failure behavior

- Gateway restart resumes queued work from the durable task ledger.
- A result committed through the exact terminal callback is replayed without
  resubmitting the turn.
- After a process crash, a local attempt whose callback identity is gone stays
  explicitly indeterminate, then defers after a bounded window so healthy Bots
  can continue. Retry starts a new fenced generation; Hermes never guesses from
  the client display transcript or silently executes the same attempt twice.
- A stale process, lease, execution generation, or cancellation generation
  cannot commit over newer work.
- Stop stays pending until the exact active attempt is terminal; completion may
  win a real race, but a later member cannot start behind an unacknowledged
  Stop.
- Dashboard and messaging-gateway processes share one fenced execution owner.

## Validation

Exact head `ee2f98efe1`, stacked directly on #97712 head `f5908c18d8`:

| Check | Result |
| --- | ---: |
| `uv run --extra dev pytest -q $(git diff --name-only origin/main...HEAD \| rg '^tests/.*\.py$' \| xargs ...)` | **221 passed** |
| Ruff on every changed Python file | passed |
| No Desktop, peer/API, RoomLink, file, artifact, messaging, or relay path in this diff | passed |
| `git diff --check` | passed |

The three review units are: runner and discussion policy; recovery and control;
bounded scheduling plus the enforced local-only boundary. The current head includes legacy-title compatibility, read-only idle polling, durable turn
deadlines, explicit crash recovery, bounded replies, non-blocking startup and
prompt fences, local-authority enforcement, and the explicit upstream
`room_plumbing` and `follow_profile_config` session contracts. Required exact-head CI is green.

This standalone head has green required CI and 221 focused tests. The complete integration build in #98307 has exercised the same-gateway runner with multiple real Bot profiles, ordered replies, files, restart, and history recovery. Files and Messaging are not part of this review delta.

## Related work

- #97681: six-layer production and field-testing plan
- #97712: required durable authority and replay parent
- #95163: gateway-hosted Group Chat design decision
- #94726 section 5: Bot Mode room lifecycle and stall class
- #94697: independent accepted-turn lifetime work. The current heads share a
  prompt-submit block; whichever lands second must rebase and preserve both
  compatible policies.
- #94570 and #91911: Stop, identity, and cancellation direction

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Security fix
- [ ] Documentation update
- [x] Tests
- [ ] Refactor

### DIFF
--- a/gateway/hosted_room_discussion.py
+++ b/gateway/hosted_room_discussion.py
@@ -1,0 +1,1320 @@
+"""Deterministic policy for same-gateway hosted-room Discussions.
+
+This module translates a frozen local member roster and the complete typed room
+log into one next driver task.  It performs no I/O, starts no workers, and knows
+nothing about transports or model runtimes.  Callers persist the returned task
+with :mod:`gateway.hosted_room_driver` and append publication plans with
+:mod:`gateway.hosted_rooms`.
+
+The unpublished driver payload intentionally remains unchanged.  Discussion
+coordinates live in deterministic ``TaskIdentity`` values and typed terminal
+events; a restart can therefore reconstruct a task without widening the driver
+schema.  Callers must reconcile terminal driver rows into publication plans
+before asking for the next task.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import re
+from collections.abc import Iterable, Mapping, Sequence
+from dataclasses import dataclass
+from typing import Any, Literal
+
+from gateway import hosted_room_driver as driver
+from gateway import hosted_rooms
+
+
+MAX_DISCUSSION_MEMBERS = 6
+MIN_DISCUSSION_MEMBERS = 2
+MAX_DISCUSSION_ROUNDS = 3
+MAX_DISCUSSION_MESSAGES = 10
+MAX_DISCUSSION_DELTA_LINES = 24
+MAX_USER_TEXT_BYTES = 64 * 1024
+MAX_MEMBER_TEXT_BYTES = 64 * 1024
+_TRUNCATED_REPLY_NOTICE = (
+    "\n\n[Reply truncated. Ask the Bot to share the full result as a file.]"
+)
+
+DecisionStatus = Literal["idle", "task", "settled", "bounded"]
+TerminalKind = Literal["settled", "failed", "cancelled", "deferred"]
+
+_IDENTIFIER_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._:-]*$")
+_MENTION_RE = re.compile(r"@([A-Za-z0-9][A-Za-z0-9._:-]*)", re.IGNORECASE)
+_TURN_ID_RE = re.compile(
+    r"^d(?P<source>[1-9][0-9]*)\.r(?P<round>[0-2])\."
+    r"p(?P<position>[0-5])\.s(?P<seen>[1-9][0-9]*)\."
+    r"m(?P<member>[0-9a-f]{24})$"
+)
+
+_MEMBER_FIELDS = frozenset({"member_id", "profile", "handle", "display_name"})
+_REMOTE_MEMBER_FIELDS = frozenset({
+    "connectionId",
+    "connectionKind",
+    "connectionLabel",
+    "connection_id",
+    "connection_kind",
+    "connection_label",
+    "remoteSource",
+    "route",
+    "sourceMissing",
+    "sourceReachable",
+    "sourceScoped",
+    "targetProfile",
+    "target_profile",
+})
+_USER_PAYLOAD_FIELDS = frozenset({"text", "thread_id"})
+_MEMBER_MESSAGE_FIELDS = frozenset({
+    "discussion_event_id",
+    "member_id",
+    "member_index",
+    "round_index",
+    "task_id",
+    "text",
+    "thread_id",
+    "turn_id",
+})
+_TERMINAL_COMMON_FIELDS = frozenset({
+    "discussion_event_id",
+    "member_id",
+    "member_index",
+    "round_index",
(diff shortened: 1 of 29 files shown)


andrexibiza (commented):
Blocking contract seam: `HostedRoomServerRPC.create()` calls `session.create` with `profile`, `title`, `source`, `hidden`, and `close_on_disconnect`, while `InternalSessionRPC.create()` and its tests have no `room_plumbing` path. Current `main`'s Desktop producer sends `room_plumbing: true` and `follow_profile_config: true`, and the server resume logic names the explicit marker as the primary contract; hidden + `Group:` is only the legacy fallback.

Thread the explicit classification through `HostedRoomRuntime -> InternalSessionRPC.create -> HostedRoomServerRPC.create -> session.create`, preserve `Group: <room_id>` for canonical migration/session reuse, and pin the actual RPC payload in `tests/tui_gateway/test_hosted_room_server_rpc.py`. The hosted path should not ship by depending on the compatibility heuristic.

dokterdok:
Addressed in current head `9560439e97`: `HostedRoomServerRPC.create()` now sends both `room_plumbing: true` and `follow_profile_config: true`, while retaining `Group: <room_id>` for migration/session reuse. The real-adapter test pins both fields in the actual `session.create` payload; the targeted hosted-room driver and adapter suite passes 75/75.

andrexibiza (commented):
Re-review of the Layer 2 delta at exact head `9560439e970ca390d1aab4feb7eb649b754da10e`, after parent head `5aead6f8750df3993c54ef3fbc20b47abefa26e2`.

The earlier session-classification blocker is repaired: the real `session.create` path now carries both `room_plumbing: true` and `follow_profile_config: true`, while retaining the canonical `Group: <room_id>` title for migration and reuse. I traced durable task admission, leases, execution/cancellation generations, terminal publication reconciliation, crash recovery, Stop settlement, policy checkpoints, local-authority selection, and the process-owned worker boundary. I found no additional Layer 2 mechanism blocker on this head. Exact-head CI, Docker, and Nix are green.

It still cannot land yet:

1. The parent #97712 replay page remains incorrectly bounded in Unicode characters rather than UTF-8 bytes. Every runner recovery and publication sweep inherits that replay boundary.
2. This delta modifies `gateway/run.py`, `hermes_cli/web_server.py`, and `tui_gateway/server.py`, all above the repository's hard 2,000-physical-line ceiling. The touched integration seams must be sharded rather than extended in place.

This head is not merge-ready until the parent blocker and touched-godfile violations are repaired, then the recomposed exact head receives fresh proof.

teknium1:
Merged via #99099 (rebase-merge, your commit and authorship preserved) — https://github.com/NousResearch/hermes-agent/pull/99099, on main at 5a3edc7467.

Two follow-ups rode on top:

- A cancel/disband race on your exact head: `HostedRoomRuntime.cancel()` trusted its initial status read, so a task transitioning `queued→running` (or settling) between the read and the state call surfaced a transient "running work requires acknowledged two-phase cancellation"/`StaleTaskError` and failed `groups.disband` (deterministic repro: `test_client_event_id_cannot_squat_disband_receipt`, 5/5 local failures on `ee2f98efe1`). `cancel()` now re-reads and re-routes on race-shaped failures with bounded retries.
- The capabilities/method registry carries both this layer's runner surface (`groups.stop/retry/approve`) and the replication surface that landed in #99047 (`groups.replicate/replica_state/promote/demote`).

Layer 3 (#97797) is next — we'll fix the silent-reauthorization blocker in the salvage per the plan on #97681.

PR #97747, opened 2026-08-29
fix(desktop): coalesce duplicate branch creates

### LINKED ISSUE #97414: [Bug]: Desktop branch re-fires on failure — duplicate child sessions accumulate and branch rows never render in the sidebar
## Bug Description

On macOS, **"Branch from here" in the Desktop sidebar creates the branch session(s) on the backend but never renders them — and the failed branch appears to re-fire automatically, accumulating duplicate child sessions.** The pane shows a blank black view with only the loading glyph (no transcript, no composer, no tab), and no branch row ever appears in the sidebar, nested or otherwise.

Two occurrences, two days apart, same machine/build:

| | Aug 26 | Aug 28 |
|---|---|---|
| User branch clicks | 1–2 | 2 |
| Child sessions created in `state.db` | **7** (all within ~70 ms of each other — single-action fan-out/retry) | **17** (bursts every ~30–90 s over 45 min — periodic re-fire, not user clicks) |
| UI result | blank pane + loading glyph | blank pane + loading glyph; **no child rows visible in sidebar at all** |

The 17 Aug 28 children all have `parent_session_id` set to the parent, 23–70 messages copied from the parent transcript, empty `title_source`, `hidden=0`, `archived=0`. So the backend side succeeds every time; the client-side resume/render step fails silently — no errors in `gui.log`, `gateway.log`, or `tui_gateway_crash.log` for any of the 17 attempts. Branching from the TUI (`/branch`) works fine on the same machine, which isolates this to the desktop branch → resume path.

## Steps to Reproduce

1. Hermes Desktop v0.17.0 (macOS 26.6.2, ARM64), packaged 2026-08-27.
2. Open a large session (parent had 663 messages) in the main pane.
3. Right-click the sidebar row → "Branch from here".
4. Observe: pane goes blank black with only the loading glyph. No branch tab, no composer, and no new row in the sidebar — not even after scrolling.
5. Over the following ~45 minutes, additional duplicate child sessions appear in `state.db` without further clicks (17 total), apparently the client re-attempting the failed branch/resume on a timer.

## Expected Behavior

One click → one branch session, which opens with the copied transcript (or at worst the loader from #96513), and the branch row appears nested under the parent in the sidebar.

## Actual Behavior

- Backend: branch RPCs succeed; each attempt creates a fully-formed draft child session with the parent transcript.
- Renderer: pane never paints; optimistic row never commits to the visible list, so the sidebar shows nothing.
- Failure is silent — zero log errors client- or gateway-side across 24 attempts over two days.
- Plausible mechanism consistent with the code: `forkBranch` creates a draft (`session.create`, no `state.db` row until first turn — by design) and relies on `resumeSession`/tile-open to paint; when that render step fails or races (parent mid-turn?), the retry path re-enters `forkBranch` and mints a *new* session each time instead of recovering the existing one.

## Additional Evidence

- `state.db` query: 17 rows `WHERE parent_session_id='<parent>'` created 14:03–14:48 on Aug 28, none titled, none hidden — invisible in the UI but fully recoverable (they resume fine when opened via other means).
- Parent transcripts intact in all cases; no data loss, just duplicates + invisible branches.
- Related but distinct: #96513 (branch takeover latches on loader until manual re-navigation — same `forkBranch`/resume chain, but no blank black pane, no invisible sidebar rows, no re-fork loop; that report is Windows), #73207 (branching branches the foreground chat), #80973 (branch drops conversation tail).

## Operating System

- macOS 26.6.2 (Apple Silicon), Hermes Desktop v0.17.0 (packaged 2026-08-27), local backend.

Ahmett101:
## Summary

Desktop branch creation could be re-entered while the first branch/resume path was still pending, so a failed or retried renderer transition could call `session.create`/`session.branch` again and accumulate duplicate child sessions. This adds an in-memory single-flight guard around branch creation keyed by the parent/source/profile/prefix, so repeated attempts reuse the existing backend child while it is in flight.

---

### `apps/desktop/src/app/session/hooks/use-session-actions/index.ts`

- Added branch-create flight tracking for Desktop branch actions.
- Coalesces duplicate create attempts onto the same pending backend child.
- Clears failed create flights so a real retry can still run.
- Preserves the existing branch open/resume paths after creation succeeds.

### `apps/desktop/src/app/session/hooks/use-session-actions.test.tsx`

- Added a regression test for concurrent duplicate stored-session branch attempts.
- Verifies only one `session.create` RPC is sent and both callers resolve successfully.

---

## How to Test

```bash
cd apps/desktop
NODE_OPTIONS=--max-old-space-size=3072 node ../../node_modules/vitest/vitest.mjs run --project ui src/app/session/hooks/use-session-actions.test.tsx
NODE_OPTIONS=--max-old-space-size=3072 node ../../node_modules/typescript/bin/tsc -p . --noEmit
node ../../node_modules/eslint/bin/eslint.js src/app/session/hooks/use-session-actions/index.ts src/app/session/hooks/use-session-actions.test.tsx
```

Result locally:

```text
Vitest: 85 passed
TypeScript: passed
ESLint targeted: passed, 0 warnings
```

---

## Checklist

- [x] Tests pass — 85/85 targeted Desktop UI tests
- [x] Follows Conventional Commits
- [x] Changes scoped to this fix only
- [x] Cross-platform impact assessed (Desktop renderer: macOS / Windows / WSL2)
- [x] profile-safe paths used — no hardcoded `~/.hermes`
- [x] `.env` not used for non-credential settings

---

## Risk & Impact

Low. This is renderer-only idempotency around branch creation; it does not change backend session schema, persistence, or the existing navigation behavior.

Type: Bug fix

Closes: #97414

### DIFF
--- a/apps/desktop/src/app/session/hooks/use-session-actions.test.tsx
+++ b/apps/desktop/src/app/session/hooks/use-session-actions.test.tsx
@@ -117,6 +117,7 @@ const RUNTIME_SESSION_ID = 'rt-new-001'
 type HarnessHandle = Pick<
   ReturnType<typeof useSessionActions>,
   | 'archiveSession'
+  | 'branchStoredSession'
   | 'createBackendSessionForSend'
   | 'openNewSessionTile'
   | 'removeSession'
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: this is the focused fix PR for #97414's duplicate-child-session branch re-entry path.

OutThisLife:
Superseded by #100879, which consolidates the full branch-load path. Your commit is cherry-picked there with authorship preserved.\n\nOne change on top: the flight key now includes the owner, since two connections can each expose a session with the same id and a route-blind key coalesced them into one create. Test in #100879 covers it.

PR #97748, opened 2026-08-29
fix(skills): a failed batch rollback never destroys the skill or its snapshot (#97714)

### LINKED ISSUE #97714: [Bug]: a failed batch rollback restore deletes the skill and its snapshots

lEWFkRAD:
## What & why
Fixes #97714.

The `operations[]` batch rollback removed the live skill directory **before** restoring from the snapshot, and the `finally` deleted the snapshot root **unconditionally**. When the restore `copytree` raised (disk full, locked file, Windows path limits), the skill directory and its only backup were both gone — the atomicity feature destroyed the skill on rollback error.

## The fix
Rollback now restores into a staging sibling (`<skill>.rollback`) first; the live directory is only swapped away once the staged copy exists, and the snapshot cleanup is skipped whenever any rollback reported `ROLLBACK FAILED`. The failure payload now names the surviving snapshot path so the operator can recover by hand.

- Restore-copy fails → live dir untouched, snapshot preserved.
- Swap fails after rmtree (rare; e.g. a Windows handle holds the dir) → the staged tree is left in place as a second recoverable copy, and the error names both paths.

## How to test
`tests/tools/test_skill_manage_batch.py::test_failed_rollback_preserves_snapshot_for_recovery` monkeypatches the module-level `copytree` to fail only the restore leg (destination under `HERMES_HOME/skills`) and asserts a `skill_batch_*` snapshot dir with the pre-batch `SKILL.md` survives on disk.

- Red on current main: `AssertionError: [] is not true : snapshot was deleted despite failed rollback`
- Green with the fix: full batch suite 9 passed; combined with `test_skill_manager_tool.py`: **69 passed**.
- Reproduced on Windows 11, Python 3.11+; `ruff check` clean.

## Platforms
Tested on Windows 11. The changed code path is pure pathlib/shutil and platform-neutral; the failure mode it guards against (locked/undeletable directories, path-length limits) is most common on Windows.

### DIFF
--- a/tests/tools/test_skill_manage_batch.py
+++ b/tests/tools/test_skill_manage_batch.py
@@ -74,6 +74,90 @@ class TestSkillManageBatch(unittest.TestCase):
         self.assertFalse(r["success"])
         self.assertFalse(os.path.exists(os.path.join(self.home, "skills", "fresh")))
 
+    def test_failed_rollback_preserves_snapshot_for_recovery(self):
+        """(#97714) A restore that raises must not destroy both copies.
+
+        The old _rollback() removed the live skill directory BEFORE copying
+        the snapshot back; if the copy then raised (disk full, locked file,
+        Windows path limits), the finally deleted the snapshot too — the
+        skill and its only backup were both gone. A failed restore must
+        leave the operator a recoverable copy, and the error payload must
+        say so instead of implying the rollback ran.
+        """
+        self._call("probe", [{"action": "create", "content": SK.format(n="probe")}])
+        smt = self.smt
+        real_copytree = shutil.copytree
+        skills_dir = os.path.join(self.home, "skills")
+        state = {"armed": False, "snap_dir": None}
+
+        real_rmtree = shutil.rmtree
+        surviving_snaps = []
+
+        def tracking_rmtree(path, *a, **k):
+            p = os.fspath(path).replace("\\", "/")
+            # record every snapshot dir that gets deleted, so the test can
+            # assert whether the backup survived
+            if "skill_batch_" in p:
+                surviving_snaps.append(p)
+            return real_rmtree(path, *a, **k)
+
+        def copytree_gated(src, dst, *a, **k):
+            dstp = os.fspath(dst).replace("\\", "/")
+            if dstp.startswith(skills_dir.replace("\\", "/")) and state["armed"]:
+                raise OSError("simulated restore failure (disk full)")
+            return real_copytree(src, dst, *a, **k)
+
+        real_skill_manage = smt.skill_manage
+
+        # The batch loop calls the module-global skill_manage per op; wrap it
+        # so the first failing op arms the restore failure (the rollback's
+        # copytree then raises on its way back into the skills dir).
+        def gated_skill_manage(*a, **k):
+            result_raw = real_skill_manage(*a, **k)
+            try:
+                ok = json.loads(result_raw).get("success")
+            except Exception:
+                ok = False
+            if not ok:
+                # this was the failing op — rollback runs next
+                state["armed"] = True
+            return result_raw
+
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #97715. Both PRs fix the failed batch-rollback restore path so a copy of the skill/snapshot survives a copy failure; #97715 is the earlier open PR.

lEWFkRAD:
Closing this PR in favor of the earlier canonical #97715 so the rollback fix has one review thread. This revision also is not safe to merge as written: `stage = Path(str(pre_dir) + '.rollback')` selects a predictable sibling name that is valid under the skill/category name grammar, and the unconditional `shutil.rmtree(stage)` can erase an unrelated legitimate `foo.rollback` skill/category. The regression test scans the process-global temp directory for any `skill_batch_*` tree, so a leftover from another test/run can make it pass; it also intentionally leaves its own snapshot behind instead of isolating, validating, and cleaning that exact snapshot. I posted the shared blocking review on #97715. Preserving this branch for reference.

kshitijk4poor:
Closing in favor of #97924 (salvage of the competing #97715 by @Adolanium, now merged).

Both PRs fixed #97714 correctly. #97715 won on the restore mechanics: its move-aside approach puts the half-applied working tree BACK when the restore copy fails, so the operator keeps a functioning skill; the stage-then-swap here can still end with only the raw snapshot on disk when the swap fails. Your PR had the better operator-facing error payload though — the "snapshot preserved at <path>" detail in the ROLLBACK FAILED note was folded into the merged fix with credit to you (commit "fix: name the preserved snapshot path in the ROLLBACK FAILED payload").

Thanks for the careful analysis — the dual-failure-mode test in this PR was genuinely thorough.

PR #97752, opened 2026-08-29
fix(agent): exclude ollama.com from deepseek reasoning_content echo-back

### LINKED ISSUE #97751: [Bug]: ollama-cloud deepseek models misclassified as DeepSeek echo-back endpoint (sends reasoning_content the endpoint discards)
## Bug Description

When using `ollama-cloud` as the provider with a deepseek-named model (`deepseek-v4-flash:0731`), Hermes classifies the endpoint as a DeepSeek echo-back family (model-name substring "deepseek" matches `_REASONING_ECHO_RULES`), so every assistant turn's full `reasoning_content` is re-sent on the next request.

ollama-cloud does **not** enforce `reasoning_content` echo-back — it ignores the field entirely. **Live experiment (2026-08-29):** replaying 0 / 10 / 1400-char `reasoning_content` all yield identical `prompt_tokens` (28.0), so the field is discarded server-side — it neither counts toward input tokens nor feeds the model. The misclassification is therefore a **correctness issue** (Hermes sends a field the endpoint ignores), not a cost issue.

Two observed symptoms (ollama-cloud + `deepseek-v4-flash:0731`, `agent.reasoning_effort: high`):
1. Final (non-reasoning) output occasionally repeats — minor
2. Reasoning chain itself repeats, cannot advance, consumes large amounts of tokens, requires manual interrupt — severe

The severe symptom is **correlated with** the echo-back misclassification but driven by a different mechanism: when the model degenerates and emits only `reasoning` with empty `content`, Hermes' `_has_structured` check (conversation_loop.py:7862) routes into thinking-prefill continuation (up to 2 full-context re-sends) then empty-response retry (up to 3 more) — each re-sending the full conversation at full price. That amplification is a separate concern (reasoning-degeneration detection) and is tracked separately from the echo-back fix in PR #97752.

## Steps to Reproduce

1. Configure `model.provider: ollama-cloud`, `model.default: deepseek-v4-flash:0731`, `agent.reasoning_effort: high`
2. Start a complex task (long context / multi-turn tool calls)
3. Observe: the reasoning chain begins repeating the same content, cannot advance, token usage keeps growing
4. Inspect the request body: assistant history messages carry a `reasoning_content` field (echo-back), while ollama-cloud actually ignores that field

Direct check of the misclassification:

```python
from agent.message_sanitization import matches_reasoning_echo_family
# Returns True (wrong — ollama-cloud does not enforce echo-back)
matches_reasoning_echo_family("deepseek", "ollama-cloud", "deepseek-v4-flash:0731", "https://ollama.com/v1")
```

## Expected Behavior

- `ollama-cloud` (ollama.com/v1) should NOT be classified as a DeepSeek echo-back endpoint; `reasoning_content` should not be re-sent for it
- `api.deepseek.com`, Kimi, and MiMo echo-back behavior unchanged

## Actual Behavior

- `matches_reasoning_echo_family("deepseek", "ollama-cloud", "deepseek-v4-flash:0731", "https://ollama.com/v1")` → `True`
- Full reasoning text re-sent every turn to an endpoint that discards it

## Root Cause

`agent/message_sanitization.py` `_REASONING_ECHO_RULES` deepseek rule matches on model-name substring "deepseek" with no host exclusion. ollama.com serves deepseek-named models but does not require echo-back.

Evidence (live, 2026-08-29):
- `deepseek-v4-flash:0731` and `qwen3.5:397b` via ollama.com/v1 both return `message.keys = ['role', 'content', 'reasoning']`; `reasoning_content` absent
- Token-cost experiment: 3 scenarios (no / short / long `reasoning_content`) → `prompt_tokens` identical (28.0) → endpoint discards the field
- Replay with `reasoning_content` in history: HTTP 200, but the model ignores the replayed reasoning (asked "what did you just think the answer was", it answered from its own new reasoning, not the replayed text)

## Affected Component

## Debug Report

Not uploaded. Reproduction is deterministic via the direct function check above; no machine config or tokens required.

## Operating System

macOS 26.3.1

## Python Version

3.11 (Hermes bundled venv)

## Additional Context

- Related but distinct: local Ollama puts thinking in `content` as `  thinking` / `  response` indented tags (no angle brackets); `strip_think_blocks` only strips angle-bracket forms, so detection (`_has_inline_thinking`) and stripping disagree. That is a separate issue (local Ollama only).
- Community search (2026-08-29): no existing issue/PR for ollama-cloud echo-back misclassification. Related reasoning-only issues: #92803 (9Router), #67321 (Codex), #83915 (truncation loop), #60928 (ollama.com misdetected as local Ollama for truncation).
- Fix: PR #97752

fred0m:
## What does this PR do?

Fixes a false positive in the reasoning_content echo-back family table: `ollama-cloud` (ollama.com/v1) serves deepseek-named models (`deepseek-v4-flash:0731`, `deepseek-v4-pro:0813`) but does **not** enforce `reasoning_content` echo-back — it ignores the field entirely (verified live against ollama.com/v1: replaying a 1400-char `reasoning_content` changes `prompt_tokens` by 0). The deepseek family rule matched on the model-name substring, so every assistant turn's full reasoning text was re-sent on the next request to an endpoint that discards it.

This is a correctness/hygiene fix: Hermes should not send a field the endpoint ignores. It also removes a source of confusion when debugging reasoning-loop issues (the replayed field is not the cause of token burn — see the experiment notes below).

The fix adds an excluded-hosts slot to `_REASONING_ECHO_RULES`; excluded hosts win over provider/model-name matches. `api.deepseek.com`, Kimi, and MiMo behavior is unchanged.

## Related Issue

Fixes #97751

## Type of Change

## Changes Made

- `agent/message_sanitization.py`:
  - `_REASONING_ECHO_RULES`: added a 6th tuple slot `excluded base_url hosts`; deepseek rule now excludes `ollama.com`
  - `matches_reasoning_echo_family()`: excluded hosts are checked first and return `False` regardless of provider/model-name matches
- `tests/agent/test_message_sanitization_policy.py`:
  - Added 2 parametrized cases: `ollama-cloud` + `deepseek-v4-flash:0731` / `deepseek-v4-pro:0813` → `None` (no echo family)

## How to Test

1. `pytest tests/agent/test_message_sanitization_policy.py -q` → 58 passed
2. Direct check (repro of the false positive):
   ```python
   from agent.message_sanitization import matches_reasoning_echo_family
   # Before fix: True (wrong — ollama-cloud does not enforce echo-back)
   # After fix:  False
   matches_reasoning_echo_family("deepseek", "ollama-cloud", "deepseek-v4-flash:0731", "https://ollama.com/v1")
   ```
3. Regression: `api.deepseek.com` / provider `deepseek` / model `deepseek/deepseek-v3` still match → `True`

## Root Cause Analysis

**Symptom (user report, ollama-cloud + deepseek-v4-flash:0731, reasoning_effort: high):**
1. Final (non-reasoning) output occasionally repeats — minor
2. Reasoning chain itself repeats, cannot advance, burns large amounts of tokens, requires manual interrupt to escape — severe

**Root cause chain (updated after live experiments, 2026-08-29):**
1. `_REASONING_ECHO_RULES` deepseek rule matches on model-name substring "deepseek" → `ollama-cloud`'s `deepseek-v4-flash:0731` is classified as a DeepSeek echo-back endpoint
2. Hermes re-sends the full `reasoning_content` on every subsequent request
3. **Experiment result:** ollama-cloud ignores the field — replaying 0 / 10 / 1400-char `reasoning_content` all yield identical `prompt_tokens` (28.0). So the echo-back does NOT inflate input tokens. It is a correctness issue (sending a field the endpoint discards), not a cost issue.
4. The severe symptom (repeated reasoning chain + token burn) is driven by a different mechanism: when the model degenerates and emits only `reasoning` with empty `content`, Hermes' `_has_structured` check (conversation_loop.py:7862) is True → thinking-prefill continuation (up to 2 full-context re-sends) → then empty-response retry (up to 3 more). Each retry re-sends the full conversation at full price. That amplification is a separate concern (reasoning-degeneration detection) and is NOT addressed by this PR.

**Evidence:**
- Live curl against ollama.com/v1 (2026-08-29): `deepseek-v4-flash:0731` and `qwen3.5:397b` both return `message.keys = ['role', 'content', 'reasoning']`; `reasoning_content` absent
- Token-cost experiment: 3 scenarios (no / short / long `reasoning_content`) → `prompt_tokens` identical (28.0) → endpoint discards the field
- `matches_reasoning_echo_family("deepseek", "ollama-cloud", "deepseek-v4-flash:0731", "https://ollama.com/v1")` → `True` before fix

**Related but distinct:** local Ollama puts thinking in `content` as `  thinking` / `  response` indented tags (no angle brackets); `strip_think_blocks` only strips angle-bracket forms, so detection (`_has_inline_thinking`) and stripping disagree. That is a separate issue (local Ollama only) and is NOT addressed by this PR.

### Code
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(agent): ...`)
- [x] I've run `pytest tests/ -q` and all tests pass (58/58 in the affected file `tests/agent/test_message_sanitization_policy.py`)
- [x] I've tested on my platform: macOS 26.3.1

- [x] I've updated relevant documentation (docstrings) — or N/A

### DIFF
--- a/agent/message_sanitization.py
+++ b/agent/message_sanitization.py
@@ -836,13 +836,13 @@ def uniquify_tool_call_ids(tool_calls: list) -> list:
 
 _REASONING_ECHO_RULES: tuple = (
     # (family, exact providers (raw), exact providers (lowered),
-    #  model substrings (lowered), base_url hosts)
+    #  model substrings (lowered), base_url hosts, excluded base_url hosts)
     ("kimi", frozenset({"kimi-coding", "kimi-coding-cn"}), frozenset(), (),
-     ("api.kimi.com", "moonshot.ai", "moonshot.cn")),
+     ("api.kimi.com", "moonshot.ai", "moonshot.cn"), ()),
     ("deepseek", frozenset(), frozenset({"deepseek"}), ("deepseek",),
-     ("api.deepseek.com",)),
+     ("api.deepseek.com",), ("ollama.com",)),
     ("mimo", frozenset(), frozenset({"xiaomi"}), ("mimo",),
-     ("api.xiaomimimo.com", "xiaomimimo.com")),
+     ("api.xiaomimimo.com", "xiaomimimo.com"), ()),
 )
 
 
@@ -864,7 +864,12 @@ def matches_reasoning_echo_family(
     """
     from utils import base_url_host_matches
 
-    _, raw_providers, lowered_providers, model_subs, hosts = _family_rule(family)
+    _, raw_providers, lowered_providers, model_subs, hosts, excluded_hosts = _family_rule(family)
+    # Excluded hosts win: an endpoint that does not enforce echo-back
+    # (e.g. ollama.com serving a deepseek-named model) must never match,
+    # regardless of provider/model name.
+    if any(base_url_host_matches(base_url, host) for host in excluded_hosts):
+        return False
     provider_lower = (provider or "").lower()
     model_lower = (model or "").lower()
     if provider in raw_providers or provider_lower in lowered_providers:
--- a/tests/agent/test_message_sanitization_policy.py
+++ b/tests/agent/test_message_sanitization_policy.py
@@ -174,6 +174,11 @@ class TestReasoningEchoFamily:
         ("DeepSeek", "whatever", "https://x", "deepseek"),
         ("openrouter", "deepseek/deepseek-v3", "https://openrouter.ai", "deepseek"),
         ("custom", None, "https://api.deepseek.com", "deepseek"),
+        # ollama-cloud serves deepseek-named models but does NOT enforce
+        # reasoning_content echo-back — excluded host must win over the
+        # model-name substring match.
+        ("ollama-cloud", "deepseek-v4-flash:0731", "https://ollama.com/v1", None),
+        ("ollama-cloud", "deepseek-v4-pro:0813", "https://ollama.com/v1", None),
         ("xiaomi", None, "https://x", "mimo"),
         ("custom", "MiMo-7B", "https://x", "mimo"),
         ("custom", None, "https://api.xiaomimimo.com/v1", "mimo"),


Enough1122:

Reviewed `agent/message_sanitization.py` + tests for PR #97752. Adds an excluded-host list to the reasoning_content echo-back family rules so `ollama.com` (which serves deepseek-named models but doesn't enforce echo-back — verified live) never matches the deepseek family, regardless of provider/model name. The excluded-hosts check runs first and returns False, which is the right precedence. Tests add the two ollama-cloud cases with `None` expectation.

Non-blocking observations:

1. **The exclusion hardcodes current ollama.com behavior.** The rationale is "verified live" that ollama.com ignores `reasoning_content`. If ollama.com later starts echoing it, this exclusion would suppress legitimate echo-back handling for deepseek-named models on ollama-cloud and silently regress to the original bug. A brief comment (or config toggle) noting the assumption and how to revert would help future maintenance.

2. **`base_url_host_matches(base_url, host)` with a None/empty `base_url`** presumably returns no match, so the exclusion is a no-op and processing continues — the `any(...)` over an empty match set is safe. Confirmed no exception path.

3. **Minor:** the rule tuple is now 6 elements (hosts + excluded_hosts) with the comment updated; the destructure in `matches_reasoning_echo_family` was updated in lockstep. Only the deepseek family has exclusions — kimi/mimo carry empty tuples, which is clean.

The fix targets a genuine false positive precisely and with a clear, testable rule. Nothing blocking.

Issue #94916, opened 2026-08-25
Kanban lifecycle honesty: dispatch --dry-run writes state and stop guard rejects review handoff


anombyte93:
## Summary

Two live Kanban lifecycle contracts disagree with the implementation:

1. `hermes kanban dispatch --dry-run` mutates durable board state by promoting eligible `blocked`/`todo` tasks and appending `promot...[truncated]

kvnloo:
AI-assisted reproduction on public main `4032a15ad0d5f55f5c97f3fa59709ca28a992543` confirms both lifecycle mismatches on Linux:

- `dispatch_once(..., dry_run=True)` changed a temporary task from `todo` to `ready` and appended a `promoted` event. The mutation occurs before the dry-run spawn check in `hermes_cli/kanban_db.py:9944-9972`.
- After a `kanban_request_review` tool call, `session_called_kanban_terminal()` returned false and `build_kanban_stop_nudge()` returned a protocol-violation nudge. `_TERMINAL_KANBAN_TOOLS` only contains complete and block in `agent/kanban_stop.py:20,50-65,69-100`.

The existing focused tests still pass (`test_dispatch_dry_run`: 1 passed; `tests/agent/test_kanban_stop.py`: 3 passed), but they do not assert either contract. I also found no open PR referencing #94916 as of 2026-08-25T20:18:00Z.

wuli666:
Opened #97753 for lifecycle mismatch **#2** only (the stop guard). Confirmed on today's `main`: `agent/kanban_stop.py`'s `_TERMINAL_KANBAN_TOOLS` is still `{"kanban_complete", "kanban_block"}`, so a worker that obeys `hermes_cli/goals.py`'s own prompt and calls `kanban_request_review` gets nudged that its card is "still `running`" (it is in `review` by then) and told to call `kanban_complete` — closing a card that must go through review. Review agents hit the same wall with `kanban_request_changes`, since the review lane spawns through the same `_default_spawn` that sets `HERMES_KANBAN_TASK`.

**Mismatch #1 (`dispatch --dry-run` writes state) is left out on purpose**, because the obvious fix makes the feature worse.

[...]

PR #97753, opened 2026-08-29
fix(kanban): treat review-handoff tools as terminal in the stop guard

wuli666:
Covers lifecycle mismatch #2 of #94916. The `dispatch --dry-run` half is deliberately **not** in this PR — see the note at the bottom.

## The bug

`agent/kanban_stop.py` gated its turn-end guard on a two-name set:

```python
_TERMINAL_KANBAN_TOOLS = frozenset({"kanban_complete", "kanban_block"})
```

`kanban_request_review` and `kanban_request_changes` also end a worker's responsibility for a card, so the guard fired at workers that had already done the right thing.

**Build worker.** `_handle_request_review` (`tools/kanban_tools.py`) calls `kb.request_review()`, which moves the card from `running` to `review`. `session_called_kanban_terminal()` then returned `False`, and the nudge told the worker:

> Task `t_...` is still `running`.
> ... Call `kanban_complete(...)` if the work is done, OR `kanban_block(...)`

Both statements are wrong at that point: the card is in `review`, and `kanban_complete` would close a card that must go through review instead. The worker either complies and closes it, or burns its nudge budget and exits as a `protocol_violation`.

This directly contradicts the prompts the same worker is running under — `hermes_cli/goals.py`'s `KANBAN_GOAL_CONTINUATION_TEMPLATE` and `KANBAN_GOAL_FINALIZE_TEMPLATE` both say *"call kanban_request_review with a summary instead"* for code changes needing same-card review.

**Review agent.** Same wall. `dispatch_once`'s review lane spawns through the same `_default_spawn`, which sets `HERMES_KANBAN_TASK`, so `kanban_stop_nudge_enabled()` is true for reviewers too — and the sdlc-review skill that lane force-loads ends its request-changes path on `kanban_request_changes`.

## The fix

Both handoff tools join the set. One frozenset, plus a comment recording why the set is "hands the card off", not "closes the card" — that is the distinction the original two names lost.

## Tests

`tests/agent/test_kanban_stop.py`:

- `test_no_nudge_after_handoff_tool` — parametrized over both tools; asserts `session_called_kanban_terminal()` is `True` and no nudge is built. Both cases fail with `agent/kanban_stop.py` reverted.
- `test_nudge_still_fires_for_non_terminal_kanban_tool` — a `kanban_comment` call still produces a nudge, so widening the set did not swallow the case the guard exists for.

`pytest tests/agent/test_kanban_stop.py tests/tools/test_kanban_tools.py tests/hermes_cli/test_kanban_review_lifecycle.py tests/hermes_cli/test_kanban_review_surfaces.py` — 65 passed, plus one failure (`test_review_tools_are_gated_and_visible_to_kanban_workers`) that reproduces identically on unmodified `main` in this environment and is unrelated to this change.

## Deliberately out of scope

- **The `--dry-run` half of #94916.** `dispatch_once` calls `recompute_ready()` unguarded, but that is what makes the preview accurate — the `ready_rows` query that follows reads `status = 'ready'`, so skipping the promotion would make `--dry-run` under-report the tasks it is being asked to preview. A correct fix needs `recompute_ready()` to compute without writing and `dispatch_once` to use the in-memory result; that is a different change. The more clearly wrong part is that `hermes_cli/kanban.py` prints `"Would promote"` for work the DB already committed. Both belong in their own PR.
- **Verifying the tool actually succeeded.** `session_called_kanban_terminal()` matches on tool name, not on result, so a *failed* terminal call already suppresses the nudge today — `kanban_complete` included, and `test_no_nudge_after_kanban_complete` pins that name-based contract with a plain `"done"` payload. This PR keeps that behaviour rather than changing it for four tools at once.

### DIFF
--- a/agent/kanban_stop.py
+++ b/agent/kanban_stop.py
@@ -1,10 +1,10 @@
 """Turn-end guard for kanban workers.
 
-Kanban workers must end with ``kanban_complete`` or ``kanban_block``. Models
-(especially GLM / Qwen families) sometimes narrate the next step
-("Let me write the report now") and stop with ``finish_reason=stop`` and no
-tool calls. Hermes treats that as a clean exit → ``rc=0`` → dispatcher
-``protocol_violation``.
+Kanban workers must end with a terminal board tool that hands the card to
+whoever owns it next. Models (especially GLM / Qwen families) sometimes
+narrate the next step ("Let me write the report now") and stop with
+``finish_reason=stop`` and no tool calls. Hermes treats that as a clean
+exit → ``rc=0`` → dispatcher ``protocol_violation``.
 
 This module is policy-only: when a kanban worker tries to finish without a
 terminal board tool, return a bounded synthetic nudge so the conversation
@@ -17,7 +17,21 @@ import os
 from typing import Any, Iterable, Optional
 
 
-_TERMINAL_KANBAN_TOOLS = frozenset({"kanban_complete", "kanban_block"})
+# Every tool that hands the card off and ends this worker's responsibility
+# for it — not just the two that close it out.  A build worker legitimately
+# finishes with ``kanban_request_review`` (goals.py's continuation and
+# finalize prompts tell it to, for code changes needing same-card review),
+# and a review agent finishes with ``kanban_request_changes`` (the
+# force-loaded sdlc-review skill's decision table tells it to).  Both move
+# the task out of ``running``, so treating them as non-terminal fired this
+# guard at a worker that had already done the right thing — nudging it to
+# call ``kanban_complete`` on a card it must not close.
+_TERMINAL_KANBAN_TOOLS = frozenset({
+    "kanban_complete",
+    "kanban_block",
+    "kanban_request_review",
+    "kanban_request_changes",
+})
 
 _DEFAULT_MAX_ATTEMPTS = 2
 
--- a/tests/agent/test_kanban_stop.py
+++ b/tests/agent/test_kanban_stop.py
@@ -88,3 +88,59 @@ def test_no_nudge_after_kanban_complete(clear_kanban_env):
 
 
 
+
+
+@pytest.mark.parametrize(
+    "tool_name,who",
+    [
+        ("kanban_request_review", "build worker handing off for same-card review"),
+        ("kanban_request_changes", "review agent sending the card back"),
+    ],
+)
+def test_no_nudge_after_handoff_tool(clear_kanban_env, tool_name, who):
+    """Handoff tools end the worker's turn just like complete/block.
+
+    Both move the card out of ``running``, and the worker is told to call
+    them — goals.py's continuation/finalize prompts name
+    ``kanban_request_review``; the force-loaded sdlc-review skill names
+    ``kanban_request_changes``. Nudging afterwards asks a worker that did
+    the right thing to close a card it must not close.
+    """
+    clear_kanban_env.setenv("HERMES_KANBAN_TASK", "t_handoff")
+    messages = [
+        {
+            "role": "assistant",
+            "content": "",
+            "tool_calls": [
+                {
+                    "id": "1",
+                    "type": "function",
+                    "function": {"name": tool_name, "arguments": "{}"},
+                }
+            ],
+        },
+        {"role": "tool", "name": tool_name, "tool_call_id": "1", "content": "ok"},
+    ]
+    assert session_called_kanban_terminal(messages) is True, who
+    assert build_kanban_stop_nudge(messages=messages) is None
+
+
+def test_nudge_still_fires_for_non_terminal_kanban_tool(clear_kanban_env):
+    """Widening the set must not swallow the case the guard exists for."""
+    clear_kanban_env.setenv("HERMES_KANBAN_TASK", "t_abc")
+    messages = [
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed `agent/kanban_stop.py` + tests for PR #97753. Widens `_TERMINAL_KANBAN_TOOLS` to include `kanban_request_review` and `kanban_request_changes`, which also hand the card off and end the worker's responsibility for it (goals.py's continuation/finalize prompts and the sdlc-review skill direct workers to call them). This stops the turn-end guard from nudging a worker that already did the right thing to `kanban_complete` a card it must not close. The docstring and comment update is accurate, and the parametrized tests cover both new tools plus a negative case (`kanban_comment` still nudges).

Non-blocking observations:

1. **The terminal-tool set is now the authoritative "worker responsibility ended" list** but lives only in this stop-guard module. The kanban lifecycle has related consumers (dispatcher lifecycle, prompts, `dispatch --dry-run` — which the PR deliberately leaves out of scope). If future handoff tools are added (e.g. an assign/reassign), keeping the set in one shared constant (imported by the guard and any lifecycle logic) would prevent drift. Worth considering as a follow-up given the note at the bottom.

2. **Scope boundary is explicit and reasonable** — the PR scopes itself to the stop guard and leaves `dispatch --dry-run` for a follow-up; the PR body documents this clearly.

3. **Minor:** several consecutive blank lines added before the new test (`\n\n\n\n`) — cosmetic only.

The fix addresses a real lifecycle mismatch (false-positive nudges on correct handoff behavior) with a minimal, well-tested change. Nothing blocking.

KeyArgo:
Corroborating #98107 (opened today, after this PR): this fix addresses it completely.

I verified the premise against current main:

- `agent/kanban_stop.py:20` — `_TERMINAL_KANBAN_TOOLS = frozenset({"kanban_complete", "kanban_block"})`, and `session_called_kanban_terminal()` (line 50) matches only against that set. A worker that correctly calls `kanban_request_review` returns `False` → the guard nudges it to "finish".
- `hermes_cli/kanban_db.py` `request_review()` (line 6501) transitions the running implementation card to the review phase and records the implementer/reviewer on the event — i.e. it genuinely ends the implementation worker's responsibility, exactly the same contract `kanban_complete`/`kanban_block` satisfy for their runs. So the terminal set omitting it is a real lifecycle gap, not a naming nit.
- Reproduces the reported failure shape: after a successful `kanban_request_review` the nudge only accepts `kanban_complete`, which fails on a card already in `review` with "unknown id or already terminal" — the worker burns nudge budget on impossible calls.

Widening the set to `kanban_request_review` + `kanban_request_changes` is the right, minimal fix, and the parametrized tests (`test_no_nudge_after_handoff_tool` over both tools + the `kanban_comment` negative) pin the exact contract this guard exists for. Confirmed the guard also matches the `role: "tool"` message name path (lines 62-65), so both the assistant-side tool_call and the tool-result path are covered. No objection; this closes the reported issue.

themezzanine:
Agreed on the direction — a build worker that hands off for same-card review has done the right thing and should not be nudged to close a card it must not close.

One case this widening lets through: the handoff is treated as terminal on the *call*, not on the board's answer. `kanban_request_review` can be refused by the ownership / goal gate, and a refused request leaves the task `running`. The test here asserts the terminal verdict with `{"role": "tool", ..., "content": "ok"}`, which the code never reads, so a rejected handoff takes the same path as an accepted one — and that is exactly the case the guard exists for: the worker still owes a terminal action, but the nudge no longer fires, the run exits `rc=0`, and the dispatcher records a `protocol_violation` on a task that is still `running`.

What I ended up with locally was to keep `kanban_complete` / `kanban_block` terminal on the assistant call as today, and treat a review handoff as terminal only on the board's success receipt for this task:

```python
def _review_handoff_landed(content: Any) -> bool:
    if isinstance(content, str):
        try:
            content = json.loads(content)
        except (TypeError, ValueError):
            return False
    if not isinstance(content, dict) or content.get("ok") is not True or content.get("error"):
        return False
    task = (os.environ.get("HERMES_KANBAN_TASK") or "").strip()
    return not task or content.get("task_id") == task
```

Two details that bit me while writing it:

- Some providers omit `name` on tool messages, so the receipt has to be correlated back to the assistant call by `tool_call_id`. Keying only on `msg.get("name")` silently misses the handoff on those providers and the guard fires on a worker that did hand off correctly.
- A `task_id` check matters: a receipt for a *different* task should not end this worker's turn.

I have the regression cases for this — successful handoff (both tools) terminal, receipt without `name` correlated by id, and still-nudges for: no receipt at all, `ok: false`, `ok: true` with an `error` riding along, a receipt for another task, and a non-JSON error string. They drop into the existing `tests/agent/test_kanban_stop.py`. Happy to push them onto this branch or open them separately — whichever is easier for you.

teknium1:
Thanks @wuli666. Your change was salvaged into #114768 with your authorship preserved (cherry-picked); #114768 — fix(kanban): stop guard no longer nudges a worker that handed its card to review (#114598, salvage #97753, #105708) — is now merged on `main` at `7e437bba91a4`, closing issues #114598. Closing this PR in favour of the landed change; if you see a case it does not cover, please open a fresh issue with the repro and tag it.

PR #97757, opened 2026-08-29
fix(config): recognize session reset policy fields

roy-gingercat-ai:
## Summary

- recognize the gateway's existing fixed-shape `session_reset` fields in `hermes config set`
- preserve the opt-in runtime shape: this does not add `session_reset` to `DEFAULT_CONFIG` or change reset behavior
- keep typo suggestions precise (for example, `session_reset.notfy` → `session_reset.notify`)
- document `session_reset.notify: false` as the supported way to keep automatic resets while suppressing user-facing reset notices

## Problem

The gateway already reads `session_reset.notify`, but the CLI validator only knew the top-level `session_reset` root. As a result, the documented command worked while printing a false warning:

```text
⚠ 'session_reset.notify' is not a recognized config key — it was saved anyway, but Hermes may not read it.
```

## Verification

- `scripts/run_tests.sh tests/hermes_cli/test_set_config_value.py tests/gateway/test_session_reset_notify.py tests/gateway/test_config.py -j 3 -q` — 163 passed
- isolated review run with `dev,messaging` extras — 174 passed
- `uv run ruff check hermes_cli/config.py tests/hermes_cli/test_set_config_value.py tests/gateway/test_session_reset_notify.py` — passed
- real CLI/runtime smoke test: `hermes config set session_reset.notify false` emitted no unknown-key warning and `load_gateway_config()` resolved `notify=False`
- `npx -y npm@11.17.0 run build:fast` — Docusaurus build succeeded (existing `/docs/llms*.txt` broken-link warnings remain)

### DIFF
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -5355,6 +5355,21 @@ _SCHEMA_DEFINED_DICT_KEYS = frozenset({
     "plugins",
 })
 
+# Optional fixed-shape sections that are intentionally absent from
+# DEFAULT_CONFIG until a user opts in. Unlike open dictionaries, their child
+# keys are runtime-defined and should still get typo detection from
+# ``hermes config set``.
+_OPTIONAL_FIXED_CONFIG_SCHEMAS = {
+    "session_reset": {
+        "mode": None,
+        "at_hour": None,
+        "idle_minutes": None,
+        "notify": None,
+        "notify_exclude_platforms": None,
+        "bg_process_max_age_hours": None,
+    },
+}
+
 # Top-level keys that can be ANY user-supplied name (platform/provider dict
 # shapes where the outer key IS user-defined).
 _DYNAMIC_TOP_LEVEL_KEYS = frozenset({
@@ -5383,6 +5398,7 @@ def _known_top_level_keys() -> set[str]:
     keys.update(_OPEN_DICT_TOP_LEVEL_KEYS)
     keys.update(_DYNAMIC_TOP_LEVEL_KEYS)
     keys.update(_SCHEMA_DEFINED_DICT_KEYS)
+    keys.update(_OPTIONAL_FIXED_CONFIG_SCHEMAS)
     return keys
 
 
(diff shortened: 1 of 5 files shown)


Enough1122:

Reviewed `hermes_cli/config.py`, tests, and docs for PR #97757. Makes the gateway's existing fixed-shape `session_reset` fields recognizable to `hermes config set` via a new `_OPTIONAL_FIXED_CONFIG_SCHEMAS` map, without adding them to `DEFAULT_CONFIG` (preserving the opt-in runtime shape). Typo suggestions stay precise (`session_reset.notfy` → `session_reset.notify`), and the drift-guard test (`_OPTIONAL_FIXED_CONFIG_SCHEMAS["session_reset"]` == `SessionResetPolicy.__dataclass_fields__`) keeps CLI validation and gateway runtime fields in lockstep. Docs updated for the new `notify` toggle.

Non-blocking observations:

1. **The drift-guard test pins *exact* set equality.** Adding a legit field to either the dataclass or the schema without the other breaks CI — which is the intent (lockstep), but it means the schema must be updated in the same change as any gateway policy field. That's fine given the test exists; just noting the coupling is now explicit and load-bearing.

2. **Schema values are all `None` placeholders.** `_validate_config_key` uses the optional schema as the traversal node, so key *existence* is validated but there's no type/shape checking for these fields. That matches the "fixed-shape, runtime-defined" framing, but a user typo in the *value* (e.g. `notify: not-bool`) won't be caught here. Consistent with how the gateway treats it; only flagging the boundary.

3. **`DEFAULT_CONFIG.get(top, _OPTIONAL_FIXED_CONFIG_SCHEMAS.get(top))`** correctly gives precedence to an actual DEFAULT_CONFIG entry over the optional schema; for a top-level key absent from both, node stays None and existing handling applies. Verified no behavior change for non-`session_reset` keys.

4. **Minor:** `_OPTIONAL_FIXED_CONFIG_SCHEMAS` docstring explains the "why" well; the nested child values being None is a reasonable sentinel choice.

A focused, well-documented config-recognition fix with a strong drift guard. Nothing blocking.

Issue #48986, opened 2026-06-19
Windows-native pytest suite has baseline failures in ACP, path, CRLF, and file-handle tests


supplefrog:
## Summary

Running the Hermes Agent Python test suite with Windows-native Python exposes multiple platform-specific baseline failures unrelated to a local code change. The failures cluster around Windows path semantics, CRLF byte counts, asyncio pipe support, and open SQLite file handles.

## Environment

- Repository: `NousResearch/hermes-agent`
- Branch/commit tested: `main` at `ef4b897a1` (`chore(release): map srojk34 author email`)
- OS/runtime: Windows host, Git Bash shell, Windows-native Python from `uv.exe`
- Python: CPython 3.11.15
- Test environment setup:
  - `uv.exe sync --extra dev --extra acp --extra messaging --extra homeassistant`
  - `uv.exe sync --all-extras` was also attempted but is blocked on Windows by `python-olm` requiring `make`/`gmake` via the `matrix` extra.

## Reproduction

From a clean `main` checkout on native Windows:

```bash
uv.exe sync --extra dev --extra acp --extra messaging --extra homeassistant
.venv/Scripts/python.exe -m pytest tests/ -q --maxfail=10 --tb=short -o 'addopts='
```

## Actual result

The run fails before completing:

```text
10 failed, 2344 passed, 9 skipped, 30 warnings in 479.11s (0:07:59)
```

The first 10 failures are:

```text
FAILED tests/acp/test_edit_approval.py::test_write_file_approval_mutates_and_request_includes_diff
FAILED tests/acp/test_ping_suppression.py::test_bare_ping_request_produces_proper_response_and_no_stderr_noise
FAILED tests/acp_adapter/test_acp_images.py::test_acp_resource_link_file_is_inlined_as_text
FAILED tests/acp_adapter/test_acp_images.py::test_acp_resource_link_image_file_is_inlined_as_image_url
FAILED tests/acp_adapter/test_acp_images.py::test_acp_resource_link_image_mime_inferred_from_suffix
FAILED tests/agent/lsp/test_workspace.py::test_normalize_path_expands_tilde
FAILED tests/agent/test_compression_concurrent_fork.py::test_review_fork_disables_compression_to_prevent_stale_parent_fork
FAILED tests/agent/test_context_references.py::test_blocks_sensitive_home_and_hermes_paths
FAILED tests/agent/test_copilot_acp_client.py::test_run_prompt_preserves_real_home_when_profile_home_available
[...]

PR #97758, opened 2026-08-29
fix(bot-mode): avoid inherited stdin on Windows

sambai-dev:
## Summary
- prevent query-file Bot Mode DM transports from inheriting Git Bash's invalid Windows stdin pseudo-handle
- apply the same `DEVNULL` contract to the policy-gated retry path
- cover both attempts with a regression test while preserving the real stream used by peer/stdin-file delivery

## Problem
On native Windows, a local `message_agent` delivery launched through Git Bash can fail before the transport child starts:

```text
warning: Making stdin inheritable failed
message_agent delivery failed: OSError: [WinError 6] The handle is invalid
```

The query-file transport never reads stdin, but `subprocess.run` implicitly inherited the shell's unusable pseudo-handle. The retry path had the same bug shape.

## Verification
- `uv run --no-sync python -m pytest tests/tools/test_bot_mode_dm.py::test_query_file_delivery_closes_stdin_for_initial_attempt_and_retry tests/tools/test_bot_mode_dm.py::test_delivery_command_round_trip_through_windows_local_shell -q --tb=short` — 2 passed
- Windows marked gate (`windows_only and not integration`) — 152 passed, 2 skipped
- Bot DM scoped suite excluding two unchanged POSIX permission assertions on Windows — 38 passed, 1 skipped, 2 deselected
- `uv run --no-sync ruff check tools/bot_mode_dm.py tests/tools/test_bot_mode_dm.py`
- `git diff --check`

## Risk
Low. Only query-file transports now receive `stdin=DEVNULL`; peer/stdin-file transports still receive their message stream unchanged.

Refs #48986

### DIFF
--- a/tests/tools/test_bot_mode_dm.py
+++ b/tests/tools/test_bot_mode_dm.py
@@ -390,6 +390,36 @@ def test_delivery_runner_preserves_child_failure_and_unlinks(tmp_path):
     assert not dm_file.exists()
 
 
+def test_query_file_delivery_closes_stdin_for_initial_attempt_and_retry(
+    tmp_path, monkeypatch
+):
+    dm_file = tmp_path / "message.txt"
+    dm_file.write_text("secret", encoding="utf-8")
+    calls = []
+    responses = [
+        subprocess.CompletedProcess([], 1, stdout="", stderr="HTTP 429 rate limit"),
+        subprocess.CompletedProcess([], 0, stdout="", stderr=""),
+    ]
+
+    def fake_run(argv, **kwargs):
+        calls.append((argv, kwargs))
+        return responses.pop(0)
+
+    monkeypatch.setattr(subprocess, "run", fake_run)
+
+    returncode = bot_mode_dm._run_delivery(
+        ["hermes", "-p", "researcher"], str(dm_file), stdin_file=False
+    )
+
+    assert returncode == 0
+    assert len(calls) == 2
+    assert [kwargs["stdin"] for _argv, kwargs in calls] == [
+        subprocess.DEVNULL,
+        subprocess.DEVNULL,
+    ]
+    assert not dm_file.exists()
+
+
 @pytest.mark.parametrize("args", [[], ["--run-delivery"], ["--run-delivery", "bad", "x"]])
 def test_delivery_main_rejects_invalid_cli(args):
     assert bot_mode_dm._delivery_main(args) == 2
--- a/tools/bot_mode_dm.py
+++ b/tools/bot_mode_dm.py
@@ -572,6 +572,7 @@ def _run_delivery(argv: list[str], dm_file: str, *, stdin_file: bool) -> int:
             proc = subprocess.run(
                 [*argv, "--query-file", dm_file],
                 check=False,
+                stdin=subprocess.DEVNULL,
                 capture_output=True,
                 text=True,
             )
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed `tools/bot_mode_dm.py` + tests for PR #97758. Fixes the inherited-stdin failure on native Windows: query-file Bot Mode DM delivery (launched through Git Bash) could fail before the transport child starts because the child inherited Git Bash's invalid Windows stdin pseudo-handle. Both the initial attempt and the policy-gated retry path now pass `stdin=subprocess.DEVNULL`, while peer/stdin-file delivery keeps its real stream. The regression test asserts both attempts use DEVNULL, verifies returncode 0 and file unlink.

Non-blocking observations:

1. **`stdin=subprocess.DEVNULL` is applied unconditionally in the query-file branch, on all platforms.** On POSIX this is a harmless no-op improvement (a query-file child has no reason to read stdin); on Windows it's the fix. Applying it everywhere keeps the code simple and avoids a platform branch — correct call.

2. **Scope is right:** only the query-file `subprocess.run` calls changed; the `stdin_file=True` path that must preserve the real peer/stdin stream is untouched. The PR body's claim matches the diff.

3. **Test detail:** `fake_run` captures `kwargs["stdin"]` for both calls and asserts both are `subprocess.DEVNULL` — a meaningful behavioral assertion (not just "doesn't crash"). The two-step response (`HTTP 429` then success) exercises the retry path end-to-end.

4. **Minor:** `capture_output=True` still pipes stdout/stderr while stdin is DEVNULL — only the stdin side changed, which is the intended scope.

Clean, minimal fix with a targeted regression test. Nothing blocking.

monerostar:

Host: Windows 11 (build 26200), Python 3.11.15, pytest 9.1.1. Live Git Bash + native `python.exe`, not a paper review. Head `25a64f39e8`.

### Live inherit failure (this is the real Windows shape)

From Git Bash with stdin closed (`exec 0<&-`), native Python `subprocess.run(..., capture_output=True)` — the query-file path without `stdin=DEVNULL` — fails before the child starts:

```
warning: Making stdin inheritable failed
INHERIT OSError winerror= 6 [WinError 6] The handle is invalid
DEVNULL rc 0
```

Same pair with `stdin=subprocess.DEVNULL` returns 0. That matches the PR: query-file transports never read stdin, but on Win32 they still inherit Git Bash's unusable pseudo-handle unless `DEVNULL` is set.

### Tests on this head

```
python -m pytest tests/tools/test_bot_mode_dm.py -o addopts= -q --tb=short
```

- 38 passed, 1 skipped, 2 failed
- Failures are the existing POSIX mode-bit assertions (`st_mode & 0o777 == 0o700`); NTFS does not enforce `0o700`. Unrelated to this diff. PR already called that out.
- Targeted: `test_query_file_delivery_closes_stdin_for_initial_attempt_and_retry`, `test_delivery_command_round_trip_through_windows_local_shell`, `test_delivery_runner_preserves_child_failure_and_unlinks` — 3 passed.

### Scope check

- Query-file initial + retry both pass `stdin=subprocess.DEVNULL`.
- Peer/`stdin_file=True` still uses the real message stream (`subprocess.run(argv, stdin=stream)`). Good — do not DEVNULL that path.
- No other open PR is doing this Bot Mode query-file inherit fix.

Looks correct and small. CI had not reported checks on this branch when I ran; the Windows evidence above is independent of that.

teknium1:
Merged via PR #99093 — your commit was cherry-picked onto current main with your authorship preserved in git log (merge commit 6cfadafb76). Also verified live on a windows-latest runner: the WinError 6 inherit failure fired on main and was clean with your fix, and the regression test passed natively. Thanks for the clean, well-tested fix, and thanks to @monerostar for the independent Win11 verification!

PR #97759, opened 2026-08-29
fix(telegram): preserve hidden text-link URLs

olehmell:
## Summary

- preserve URL targets stored in Telegram text_link entities
- handle both message text and media captions while deduplicating visible or repeated URLs
- retain enriched captions through Telegram media handling

## Why

Telegram forwarded posts can display labels such as Article or Comments while storing the actual destination only in MessageEntity.url. Hermes previously forwarded only the visible label, so the agent received no actionable URL.

## Tests

- scripts/run_tests.sh tests/gateway/test_telegram_channel_posts.py -q
- scripts/run_tests.sh tests/gateway/test_telegram_channel_posts.py tests/gateway/test_telegram_documents.py tests/gateway/test_telegram_audio_vs_voice.py tests/gateway/test_telegram_group_gating.py -q
- venv/bin/ruff check plugins/platforms/telegram/adapter.py tests/gateway/test_telegram_channel_posts.py

Result: 46 relevant tests passed; Ruff passed.

## Platform

- Telegram
- macOS

### DIFF
--- a/plugins/platforms/telegram/adapter.py
+++ b/plugins/platforms/telegram/adapter.py
@@ -9972,7 +9972,7 @@ class TelegramAdapter(BasePlatformAdapter):
                 _observe_type = self._media_message_type(_m)
                 _event = self._build_message_event(_m, _observe_type, update_id=update.update_id)
                 if _m.caption:
-                    _event.text = self._clean_bot_trigger_text(_m.caption)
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: open maintainer salvage #93346 already addresses hidden text_link URLs with UTF-16-safe inline expansion. This PR appends deduplicated targets instead; maintainers should choose one implementation.

PR #64763, opened 2026-07-15
fix(signal): correct heading/code-block formatting interactions (clean rebuild of #52958)

Bartok9:
Clean rebuild of #52958 against current `main` — the original branch was thousands of commits behind and un-rebasable (bundled unrelated test-runner + website history).

Per the hermes-sweeper review on #52958, this PR isolates **only** the Signal-formatter commit `7dd0e1374de4` and its formatter tests; the unrelated `scripts/run_tests_parallel.py` runner changes are dropped and can be handled separately against current main.

### Fix
`gateway/platforms/signal_format.py` recorded fenced-code spans before the heading pass but then removed heading markers (`# `, `## `, ...) without re-basing style offsets, so a `#`-prefixed line **inside** a fenced code block (shell/Python comment, Markdown sample) was mis-detected as a heading — its marker stripped and bolded, corrupting the code. The heading pass now skips any offset already claimed by a MONOSPACE code span, and heading-marker removals are recorded so MONOSPACE spans are correctly re-based onto the marker-stripped output.

### Validation on current main
- Premise confirmed: `signal_format.py:50-55` record code spans, `:60-72` strip headings, `:84` inline match — all still present on `main`.
- `tests/gateway/test_signal_format.py`: **57 passed** against current main.

Full credit to the original analysis in #52958.

PR #52958, opened 2026-06-26
fix(signal): correct heading/code-block formatting interactions

Bartok9:
## Summary

Three related Markdown→Signal `bodyRanges` bugs in `markdown_to_signal()`, all in how the heading and inline passes interact with code-block bodies that were extracted to plain text earlier:

1. A code block **following a heading** lost (or misplaced) its MONOSPACE formatting.
2. A `#`-prefixed line **inside a fenced code block** was mistaken for a heading — marker stripped + spurious BOLD — corrupting the code body.
3. Inline markers (`**`, `__`, `~~`, `_`) **inside a fenced code block** were treated as formatting, stripping markers from code like `a ** b ** c`.

## Root Cause

In `gateway/platforms/signal_format.py`, fenced code blocks are extracted to plain text FIRST and recorded as MONOSPACE spans. Later passes (heading marker removal, then inline `**`/`_`/`~~` matching) operate on that flattened text and did not account for the code-block spans:

- **Bug 1:** heading-marker removal (`"# "`) shifts following text left, but the code-block spans captured earlier were never re-based, so the bounds check `cp_start + cp_len > len(text)` silently dropped the MONOSPACE span. Inline styles escaped this because their removals are tracked via `removals`/`_adjust`; heading removal had no equivalent.
- **Bug 2:** a `#` line that used to be inside a fence is indistinguishable from a real heading to `^#{1,6}\s+` once the fence is gone.
- **Bug 3:** the inline pattern loop ran over the whole text without seeding the `occupied` set with the code-block spans, so `**`/`_` pairs inside code matched.

**Evidence (captured before the fix):**
```
>>> markdown_to_signal('# Title\n\n```\ncode here\n```')
('Title\n\ncode here', ['0:5:BOLD'])                       # MONOSPACE dropped
>>> markdown_to_signal('```\n# not a heading\ncode\n```')
('not a heading\ncode', ['0:13:BOLD', '0:18:MONOSPACE'])   # '#' eaten + bogus BOLD
>>> markdown_to_signal('```\na ** b ** c\n```')
('a  b  c', ['0:7:MONOSPACE', '2:3:BOLD'])                 # code corrupted
```

[...]

PR #97761, opened 2026-08-29
fix(signal): don't let code blocks be re-mangled by heading/inline markdown regexes

xu-kai-quan:
## What does this PR do?

`markdown_to_signal()` (`gateway/platforms/signal_format.py`) de-fences
\`\`\`\`\`\`\` code blocks into plain text early — recording only a MONOSPACE
style range for them — before running the heading/bold/italic/strikethrough
passes over that same text. Those later passes have no way to know a given
span came from a code block, so real code gets corrupted:

1. A code comment line starting with `#` (e.g. Python) is read as a
   Markdown heading: the `#` is stripped from the message text and a
   spurious BOLD span is layered on top of MONOSPACE.
2. Underscore/asterisk-wrapped code (e.g. `` _var_ ``) loses its literal
   underscores/asterisks to the italic/bold passes.
3. Separately: if a heading appears *earlier* in the same message, the
   heading pass shortens `text` without adjusting the already-recorded
   code-block style position — once that stale position runs past the
   new (shorter) `len(text)`, the final bounds check silently drops the
   MONOSPACE style for the whole code block.

Net effect: sending almost any real code snippet with a comment or an
underscore over Signal silently corrupts the message content.

## Related Issue

No existing issue — found via manual code audit + repro scripts (see below).

## Type of Change

## Changes Made

- `gateway/platforms/signal_format.py`:
  - Track `code_ranges` alongside the MONOSPACE styles recorded for each
    de-fenced code block.
  - Skip heading matches that fall inside a code range.
  - Seed the inline-pattern pass's `occupied` list with `code_ranges` so
    bold/italic/strikethrough/backtick matching skips code text too.
  - Re-adjust code-block style positions after the heading pass (which
    shortens `text` by stripping `#+ ` markers), mirroring the existing
    `removals`/`_adjust` pattern already used later for the inline pass.
- `tests/gateway/test_signal_format.py`: 3 new regression tests.

## How to Test

```python
from gateway.platforms.signal_format import markdown_to_signal

# Bug 1: comment inside a code block
markdown_to_signal("```python\n# a comment\nx = 1\n```")
# before: '# a comment' -> '# comment' loses its '#', gets a spurious BOLD
# after:  '# a comment' preserved verbatim, only MONOSPACE

# Bug 2: heading before a code block
markdown_to_signal("# Heading\n\n```\ncode here\n```")
# before: styles == ['0:7:BOLD']              (MONOSPACE silently dropped)
# after:  styles == ['0:7:BOLD', '9:9:MONOSPACE']
```

`pytest tests/gateway/test_signal_format.py -q` → 23 passed (20 existing +
3 new), no regressions. `scripts/check-windows-footguns.py
gateway/platforms/signal_format.py` clean.

## Checklist

- [x] I've run `pytest tests/gateway/test_signal_format.py -q` and it passes
- [x] I've added tests for this change
- [ ] Platform tested: Windows 11 (dev environment); logic is pure string/regex handling, no I/O, OS-independent

### DIFF
--- a/gateway/platforms/signal_format.py
+++ b/gateway/platforms/signal_format.py
@@ -48,16 +48,27 @@ def markdown_to_signal(text: str) -> tuple[str, list[str]]:
     styles: list[tuple[int, int, str]] = []
 
     code_block = re.compile(r"```[a-zA-Z0-9_+-]*\n?(.*?)```", re.DOTALL)
+    code_ranges: list[tuple[int, int]] = []
     while match := code_block.search(text):
         inner = match.group(1).rstrip("\n")
         start = match.start()
         text = text[: match.start()] + inner + text[match.end() :]
         styles.append((start, len(inner), "MONOSPACE"))
+        code_ranges.append((start, start + len(inner)))
 
+    # Once de-fenced, a code block's own text (e.g. a Python "# comment" or
+    # "**kwargs") is indistinguishable from prose to the regexes below.
+    # Keep it out of the heading pass and seed `occupied` (below) with it so
+    # the inline patterns skip it too — otherwise code content gets mangled
+    # (its "#"/"**"/"_" markers stripped) and double-styled on top of MONOSPACE.
     heading = re.compile(r"^#{1,6}\s+", re.MULTILINE)
     new_text = ""
     last_end = 0
+    num_code_styles = len(styles)
+    heading_removals: list[tuple[int, int]] = []
     for match in heading.finditer(text):
+        if any(cs <= match.start() < ce for cs, ce in code_ranges):
+            continue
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #64763: both repair Signal fenced-code protection from heading/inline markdown passes and rebase code-style offsets after heading removal. #64763 is the earlier open clean rebuild.

xu-kai-quan:
Confirmed duplicate — closing in favor of #64763 (clean rebuild of #52958), which already fixes this exact interaction (code-block spans vs. the heading pass) and is further along (64 additions, 57 passing tests vs. my 56/23). Thanks for the triage flag.

PR #97766, opened 2026-08-29
fix(agent): preserve productive continuations and force overflow compaction

sgtworkman:
## Summary
- replace the fixed four-attempt non-tool continuation ceiling with a progress-based guard
- stop only after three consecutive empty or duplicate fragments while preserving session cleanup
- force compaction on provider-proven overflow so automatic cooldown cannot consume the retry budget without changing the transcript
- add regressions for continuation beyond four fragments, duplicate-fragment cleanup, next-turn usability, and forced overflow recovery

## Verification
- `13 passed` across continuation and forced-overflow focused tests
- `16 passed` across output-cap, context overflow, 413, and compression-lock regression coverage
- `git diff --check`

### DIFF
--- a/agent/conversation_loop.py
+++ b/agent/conversation_loop.py
@@ -16,6 +16,7 @@ resolved through :func:`_ra` so those patches keep working.
 
 from __future__ import annotations
 
+import hashlib
 import json
 import logging
 import os
@@ -1964,6 +1965,8 @@ def run_conversation(
     failed = False
     codex_ack_continuations = 0
     length_continue_retries = 0
+    length_no_progress_retries = 0
+    length_fragment_fingerprints = set()
     # Total outer-loop exceptions this turn (#92450) — see _MAX_OUTER_LOOP_ERRORS.
     _outer_error_count = 0
     truncated_tool_call_retries = 0
@@ -3917,6 +3920,8 @@ def run_conversation(
                                         _frag.pop("_length_continuation_nudge", None)
                                 agent._session_messages = messages
                                 length_continue_retries = 0
+                                length_no_progress_retries = 0
+                                length_fragment_fingerprints.clear()
                                 truncated_response_parts = []
                                 retry_count = 0
                                 compression_attempts = 0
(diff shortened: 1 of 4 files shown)


Enough1122:

Reviewed `agent/conversation_loop.py` + tests for PR #97766. Two related changes: (1) replaces the fixed four-attempt continuation ceiling with a progress-based guard — stop only after three consecutive empty/duplicate fragments, fingerprinting each fragment by whitespace-normalized sha256 — and (2) adds `force=True` to `_compress_context` at the provider-proven overflow sites so automatic cooldown can't consume the retry budget without changing the transcript. Tests cover continuation beyond four fragments, duplicate-fragment cleanup, next-turn usability, and forced-overflow recovery.

Non-blocking observations:

1. **The fixed 4-attempt ceiling is replaced by a no-progress trigger with no absolute cap.** As long as each fragment is non-empty and *unique* (even trivially — e.g. a provider emitting `"a"`, `"ab"`, `"abc"`, … or churning out distinct tokens forever), `length_no_progress_retries` stays 0 and the loop keeps issuing continuation API calls until the outer per-turn budget (`max_iterations`) or an interrupt stops it. The set-based fingerprint does correctly catch alternation (A,B,A,B repeats an earlier fingerprint → no-progress → stop after 3), so the only unbounded case is endless *genuinely new* fragments. Consider an absolute continuation-attempt cap (even a generous one) on top of the no-progress guard to bound pathological burn-through of the iteration budget.

2. **Fingerprinting is content-exact, not progress-measured.** A fragment that only *slightly* advances (e.g. adds a few tokens to a stalled draft) counts as progress and resets the counter — reasonable, but it means the guard measures "new text appeared" rather than "meaningful progress". Acceptable proxy; just noting the semantics for anyone tuning the threshold.

3. **`force=True` at 4 sites.** The test covers one (output-cap overflow). The other three hunks sit in adjacent overflow/compression branches — worth a quick audit that each is genuinely *provider-proven* overflow (provider returned a concrete error) and not a heuristic-based compression, so forcing doesn't bypass a legitimate cooldown where the provider never actually rejected anything.

4. **Minor:** `length_continue_retries` appears to now be display-only (the `(attempt {length_continue_retries})` messages) while the gate uses the new counter; if it's no longer incremented anywhere, the display would pin at 0 — the diff doesn't show the increment line, so confirm it's either still incremented or the message is updated to the real attempt count.

The direction (progress-based instead of ceiling-based) fixes a real over-eager-stopping bug, and the overflow-force change is sensible; the main hardening ask is an absolute cap (observation 1). Nothing blocking.

PR #97767, opened 2026-08-29
fix(photon): persist inbound replay dedupe across restarts

tachyon-r:
## Summary
- persist Photon inbound message claims in profile-scoped state
- record IDs only after successful dispatch so failures remain retryable
- validate, expire, and bound restored dedupe entries
- add restart, profile-isolation, capacity, and failed-dispatch coverage

## Verification
- all 132 Photon plugin tests passed
- Ruff, ty, Windows-footgun, and diff checks passed

### DIFF
--- a/plugins/platforms/photon/adapter.py
+++ b/plugins/platforms/photon/adapter.py
@@ -11,6 +11,7 @@ import asyncio
 import base64
 import json
 import logging
+import math
 import os
 import re
 import secrets
@@ -41,6 +42,7 @@ from gateway.platforms._shared import get_scoped_secret as _get_scoped_secret
 from gateway.platforms.base import BasePlatformAdapter, SendResult
 from gateway.platforms.event import MessageEvent, MessageType
 from gateway.platforms.helpers import compile_mention_patterns, strip_markdown
+from utils import atomic_json_write
 
 from .auth import load_project_credentials
 # Sidecar dir resolution is lazy (never at import): it probes the filesystem and may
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed `plugins/platforms/photon/adapter.py` + tests for PR #97767. Making the at-least-once inbound dedupe durable is well-motivated (a sidecar/gateway restart previously forgot dispatched IDs), and the design is sound: the 48h/4000-ID window now lives under `HERMES_HOME/state`, is loaded with strict validation (finite timestamps, wall-clock rollback clamp, newest-bounded window), written atomically with mode `0600`, and persisted only after dispatch succeeds — so a failed dispatch is still retried after restart (covered by `test_failed_dispatch_is_retried_after_restart`). The profile-scoping and bound-eviction tests are solid.

Non-blocking observations:

1. **Full-file persist on every inbound message.** `_mark_seen_message()` calls `_persist_seen_messages()` on each new ID, which serializes and atomically rewrites up to 4,000 entries. On a high-throughput Photon channel this is one disk write per dispatched message. A debounce/throttle (e.g. flush at most every N seconds or N messages, plus on shutdown) would cut the write amplification without changing correctness — the in-memory dict already provides in-process replay protection.
2. **Crash between dispatch success and persist re-dispatches the replay.** That is the correct at-least-once trade-off (and the point of persisting after success), but it means the durability guarantee is "last successful dispatch may replay once" — worth stating in the docstring so operators don't expect exactly-once after a hard kill.
3. **File grows to 4,000 lines under `indent=0` + `separators`.** The state file is one JSON document regardless, so formatting is cosmetic only — not an issue, just noting it's the full bounded window on disk every flush.
4. **Minor:** `from hermes_constants import get_hermes_home` is a local import inside `__init__`; the other modules appear to import at top level. Local import is defensible here (keeps `adapter.py` import graph light), just inconsistent with the file's style.

The core fix is correct and well-tested; nothing blocking.

tachyon-r:
Rebased onto bc1330eebc0aa8a443501b5f62586eb7361353a5 and pushed 579bce12f19213f85a4d992a3634b1d1fe3cc954 with an explicit original-head lease. The restart replay regression still fails on that base; scripts/run_tests.sh tests/plugins/platforms/photon/ passes 136 tests on the PR. Added the requested docstring documenting the dispatch-to-persist crash window: this is bounded at-least-once replay suppression, not exactly-once delivery. Kept synchronous persistence rather than widening the crash-loss window through debouncing.

Fresh CI passes unit, macOS, Windows, lint, Docker and Nix checks. E2E run 34024983418 failed only test_openai_stream_final_tool_call_delta_reaches_relay_parent_event at tests/e2e/test_relay_native_openai_stream.py:76: relay_finalizer_started.wait(5), with reciprocal finalizer wait timeout. This is outside the Photon diff; no unrelated timeout changes made. Both failed-run and individual-job rerun requests were denied by GitHub permissions. Local e2e reproduction is blocked by missing aiohttp in the existing venv; no dependencies installed. Maintainer rerun requested.

PR #68805, opened 2026-07-21
fix(state): loud failed state.db snapshot + zeroed-file quarantine

smfworks:
## Summary

Hardening for [#68474](https://github.com/NousResearch/hermes-agent/issues/68474) while the storage-stack root cause remains under investigation.

A zeroed `state.db` (full original size, all NULs, empty WAL) was observed after a Windows desktop update. Pre-update snapshot already uses `_safe_copy_db`, but a **failed** copy of a present `state.db` was only a logger warning — the update could still print success with exit 0. Opening the zeroed file later produced a generic "file is not a database" with no recovery guidance.

## Changes

1. **Loud failed snapshot** (`hermes_cli/backup.py`)
   - Track `failed_dbs` when `_safe_copy_db` fails for a present `*.db`
   - Print critical stdout warnings (not log-only)
   - Detect zeroed signature during snapshot failure
   - Record `failed_dbs` in `manifest.json`
   - If nothing else was captured, abort snapshot with an explicit message

2. **Zeroed signature at SessionDB open** (`hermes_state.py`)
   - `is_zeroed_state_db()` / `quarantine_zeroed_state_db()`
   - On open: if size>0 and header is all-NUL (no `SQLite format 3`), **quarantine** the file (preserve bytes as `state.db.zeroed-<ts>.bak`), log recovery guidance pointing at `state-snapshots` / `hermes snapshot list|restore`, then open a fresh DB so the agent can start

## Out of scope

Root cause of the zeroing event itself (may be below SQLite / storage stack). This PR only makes failure **loud** and recovery **guided**.

## Test plan

- [x] `is_zeroed_sqlite_file` true on all-NUL file, false on valid DB / empty
- [x] failed `_safe_copy_db` during quick snapshot prints CRITICAL and records `failed_dbs`
- [x] SessionDB quarantines zeroed file and opens fresh schema
- [x] Existing safe-copy + snapshot tests still green

Related: #68474

PR #97768, opened 2026-08-29
fix(backup): bound incomplete snapshots without losing DB recovery copies

tachyon-r:
## Summary
- bound retention of repeated incomplete quick snapshots
- preserve the newest recovery copy of every database omitted from the current snapshot
- handle alternating failures where no single older snapshot is complete
- add regression coverage for stale partial snapshots and per-database recovery

## Verification
- 86 focused backup tests passed; 1 skipped
- Ruff and diff checks passed

### DIFF
--- a/hermes_cli/backup.py
+++ b/hermes_cli/backup.py
@@ -1112,8 +1112,7 @@ def _copy_quick_snapshot_files(
     """Copy every quick-snapshot candidate into *staging_dir*.
 
     Returns ``(manifest {rel: size}, failed_dbs, oversized_skipped)``. The last two are snapshot
-    incompleteness (#68805): the caller must suppress pruning so the older snapshot that may hold
-    the only recoverable DB survives.
+    incompleteness (#68805): pruning still runs but must keep the latest complete recovery source.
     """
     manifest: Dict[str, int] = {}
     failed_dbs: list[str] = []
@@ -1193,18 +1192,20 @@ def _create_quick_snapshot_locked(
         json.dump(meta, f, indent=2)
     os.replace(staging_dir, root / snap_id)
     # Auto-prune (pre-update callers pass a smaller keep so state.db copies don't accumulate).
-    # Skip when a DB failed to capture OR was skipped for size (#68805): the snapshot is
-    # incomplete and the older one may hold the only recoverable database.
-    if not (failed_dbs or oversized_skipped):
-        _prune_oldest(_snapshot_dirs(root), _QUICK_DEFAULT_KEEP if keep is None else keep, shutil.rmtree, "snapshot")
-    else:
+    # When a present DB failed to capture OR was skipped for size (#68805), preserve the latest
+    # complete recovery source in addition to the retention window. Still prune stale incomplete
+    # snapshots: they contain config/auth copies too, and repeated failures must stay bounded.
+    incomplete = failed_dbs or oversized_skipped
+    retention = _QUICK_DEFAULT_KEEP if keep is None else keep
+    if incomplete:
         if oversized_skipped:
-            print("  ⚠ Skipping snapshot prune: DB file(s) skipped for size: " + ", ".join(oversized_skipped))
+            print("  ⚠ Preserving latest complete snapshot: DB file(s) skipped for size: " + ", ".join(oversized_skipped))
             logger.warning("Quick snapshot skipped oversized DB file(s): %s", ", ".join(oversized_skipped))
         logger.warning(
-            "Skipping snapshot prune because %d DB(s) failed to capture and/or %d were oversized "
-            "— preserving older snapshots as recovery source",
+            "Snapshot incomplete because %d DB(s) failed to capture and/or %d were oversized "
+            "— preserving latest complete recovery source",
             len(failed_dbs), len(oversized_skipped))
+    _prune_quick_snapshots(root, keep=retention, preserve_latest_complete=bool(incomplete))
     logger.info("quick snapshot phase=copy status=complete id=%s files=%d bytes=%d",
                 snap_id, len(manifest), sum(manifest.values()))
     return snap_id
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed `hermes_cli/backup.py` + tests for PR #97768. The change is a good correction to the #68805 behavior: instead of skipping pruning entirely whenever a snapshot is incomplete (which let stale secret-bearing partial snapshots accumulate unbounded), it always prunes to the keep window and adds one safety exception — preserve the newest complete snapshot as a recovery source. `preserve_latest_complete` scans manifests newest-first and retains the first snapshot with no `failed_dbs`/`oversized_skipped`. Tests cover both incomplete kinds (oversized/failed) and assert stale partials are removed while the complete recovery copy survives.

Non-blocking observations:

1. **The "preserve newest complete" scan ignores the keep window.** `retained = set(dirs[:keep])` then the scan adds the newest complete snapshot *regardless of age*. If the newest complete snapshot is already inside the kept window, the scan re-adds the same dir (harmless). But if a complete snapshot sits just outside the window, retention is effectively `keep + 1` even when the newer kept snapshots are also complete — slightly more aggressive retention than the intent ("when the retained window contains no complete snapshot"). Consider skipping the scan when any of `dirs[:keep]` is already complete, or documenting that `keep` is a floor.
2. **A "complete" snapshot may still be stale.** The preserved recovery copy can be arbitrarily old (state.db from weeks ago) — that is the stated intent (a recovery source), but the chosen newest complete may be much older than the user expects if failures persist. The warning message ("preserving latest complete recovery source") is accurate; just flagging the semantics for the message consumers.
3. **Corrupt manifest falls through silently.** A snapshot with an unreadable/corrupt `manifest.json` is skipped by the scan (never counted complete), so it can be pruned even if it contains the only good DB copy. Given manifests are written by the same tool this is unlikely, but a debug log on manifest parse failure would help if it ever happens.
4. **Minor:** the existing `test_oversized_db_suppresses_pruning` still passes (via the "skipping state.db" assertion), but its name and the removed "Skipping snapshot prune" message no longer reflect the new always-prune behavior — worth renaming/updating to assert the new "preserving latest complete" contract.

Solid fix with good test coverage; nothing blocking.

daerias:
**Review: `fix(backup): bound incomplete snapshot retention`**

**Summary**

Thanks for this — bounding retention while keeping a recovery source is exactly the right direction after #68805. I checked out the branch and ran the backup tests locally: **59/59 passed** (`tests/hermes_cli/test_backup.py`). Treating oversized skips as incompleteness (`hermes_cli/backup.py:1675-1676`) and preserving the latest complete snapshot instead of suppressing the prune (`backup.py:1690-1711`, `2054-2068`) is sound, and the new parametrized test is exactly the regression coverage this needs.

Three findings; the first is the one I'd want to see addressed.

**1. `preserve_latest_complete` pins only fully clean snapshots — the last copy of a DB can still be pruned** *(substantive)*

The completeness check at `backup.py:2062-2066` requires both `failed_dbs` and `oversized_skipped` to be empty. A snapshot that is itself partial — but holds the newest surviving copy of a *different* DB — is never pinned, and the prune loop (`backup.py:2071-2076`) deletes it.

<details>
<summary>Repro (production paths only)</summary>

1. `_QUICK_STATE_FILES` (`backup.py:1434-1443`) includes both `state.db` and `cron/executions.db`.
2. Snapshot S1: `state.db` copies fine, `cron/executions.db` exceeds the cap → S1's manifest records `oversized_skipped: ["cron/executions.db"]` (`backup.py:1612-1614`). S1 is "incomplete" — yet holds the newest `state.db` copy.
3. Snapshot S2 (later, `keep=1`): `state.db` exceeds the cap → `oversized_skipped: ["state.db"]`. Also incomplete.
4. With `keep=1`, `retained = {S2}` (`backup.py:2054`). The preserve loop finds no snapshot with *both* lists empty and pins nothing. S1 is pruned → the last `state.db` copy is gone.

</details>

**Impact:** silent loss of the recovery copy — the failure mode this PR prevents, one step removed.

**Suggested fix:** preserve granularly, per missing DB — pin the newest older snapshot whose `files` contains each DB in the current snapshot's `failed_dbs` + `oversized_skipped`. Retention stays bounded (one pinned dir per missing DB); stale partials holding no missing DB are still pruned. I've built this as a RED/GREEN pair locally: a reproducer test (RED against this branch) plus a granular-preserve patch (GREEN; `test_oversized_db_suppresses_pruning` and the new parametrized test stay green). If you prefer keeping the simpler all-or-nothing rule, documenting this edge as a known limitation is also reasonable — happy to attach the patch here or open a follow-up PR, your call.

**2. `test_oversized_db_suppresses_pruning` no longer describes what it tests** *(rename)*

`tests/hermes_cli/test_backup.py:1355`. Pruning is no longer suppressed — it proceeds and the complete snapshot is preserved instead. The assertion still guards the right invariant, but the name/docstring describe the pre-PR behavior. Suggest `test_oversized_db_preserves_latest_complete_recovery_snapshot`.

**3. Manifest read failures are silently swallowed** *(logging)*

`backup.py:2060-2061` — `except (OSError, json.JSONDecodeError): continue`. A corrupt manifest makes the preserve scan skip silently; the prune loop then deletes that snapshot with no signal (the prune path itself logs failures, `backup.py:2077-2078`). Suggest at least a `logger.warning` in the except branch.

<details>
<summary>Nitpick (optional)</summary>

`backup.py:2064-2065` classifies via `.get("failed_dbs")` / `.get("oversized_skipped")` — manifests written before the `oversized_skipped` field existed read as "complete", so a legacy snapshot that actually skipped a DB could be pinned as the recovery source. Minimal impact, but a `manifest_version` or explicit key-presence check would make it airtight.

</details>

**Overall**

Nice fix — bounded retention is the right direction, and Finding 1 aside the implementation is clean and well-tested. Happy to contribute the granular patch directly. Thanks for keeping the recovery copy safe!

daerias (commented):
All three findings addressed — thanks for the quick turnaround.

1. **Granular recovery preservation** — `_prune_quick_snapshots` now takes `preserve_recovery_for` and pins the newest older snapshot whose manifest contains each omitted DB. That closes the silent-loss edge case where the last copy of one DB could still be pruned.
2. **Test renamed** — `test_oversized_db_preserves_latest_recovery_copy` now describes the invariant it guards.
3. **Manifest read failures logged** — the `except` branch emits a `logger.warning` instead of silently continuing.

CI is green and the diff reads clean. LGTM — approving.

PR #95940, opened 2026-08-27
search_files no longer blows up memory/transport on giant single-line files (port cline#13525)

teknium1:
## Summary

`search_files` no longer floods the exec transport and Python memory when a match lands inside a giant single-line file (serialized trace dumps, minified bundles). One match in a 40MB single-line JSON crossed the transport as a 42MB stdout payload and drove ~180MB of peak Python allocation before the per-match `[:500]` clamp ran — `head -n` counts lines, so a single enormous matched line sailed through untruncated.

Ported from [cline/cline#13525](https://github.com/cline/cline/pull/13525), where the same class crashed their CLI/hub daemon with `RangeError: Out of memory`. Hermes buffers in Python, so our failure mode is transport/memory blowup (worse on remote backends where the payload crosses the wire) rather than a hard crash — same class, fixed at the same layer.

## Changes

- `tools/file_operations.py` `_search_with_rg`: add `--max-columns 2000 --max-columns-preview` for content-mode searches — rg truncates the printed line at the engine layer; `--max-columns-preview` keeps a truncated prefix so the match is still reported instead of omitted.
- `tools/file_operations.py` `_search_with_grep` + `_search_with_grep_pruned` (darwin pruned fallback): grep has no `--max-columns`, so bound the pipeline with `| cut -c1-2000`.
- `files_only` / `count` modes are skipped in all three sites (their lines are paths/counts, never giant).
- 2000 cols exceeds the existing 500-char per-match content clamp, so nothing previously visible changes.
- `tests/tools/test_search_giant_line_containment.py`: real-pipeline tests (actual rg/grep via bash, 5MB single-line fixture) asserting the giant match is still reported, content ≤500 chars, and total stdout stays under 1MB; plus files_only/count regression coverage. Sabotage-verified: both engine tests fail on origin/main without the fix.

## Validation

| | Before (origin/main) | After |
|---|---|---|
| stdout crossing transport (1 match in 40MB single-line file) | 41.9 MB | 0.0 MB |
| peak Python alloc during search | 179.5 MB | 11.4 MB |
| matches returned | 2 (content clamped to 500) | 2 (identical) |
| targeted tests | — | 130 passed, 2 skipped (pre-existing darwin skips) |

**Live repro:** scripted `ShellFileOperations.search()` against a real 40MB single-line fixture on origin/main (numbers above), re-run on the branch to confirm containment; grep fallback exercised separately by forcing `_has_command`.

## Infographic

Search tool giant line containment

PR #95160, opened 2026-08-26
perf(file-ops): collapse read_file/write_file shell probes into one round-trip

yaojiejia:
## What does this PR do?

`ShellFileOperations.read_file` (`tools/file_operations.py`) currently issues four separate `env.execute()` calls per read: size probe, `head -c 1000 | base64` sample, `sed -n | cut` page, and `wc -l`, plus a fifth `tail -c 1 | wc -l` whenever the page reaches EOF. `write_file` issues four (line-ending probe, BOM probe, atomic write, `sha256sum`). Each call is a bash spawn locally and a full network round-trip on SSH, Docker, Daytona and Modal backends.

The probes accumulated between Aug 6 and Aug 10 as four independent correctness fixes (`e40315d53`, `e0b500598`, `faa188bf5`, `fd452e26e`). Each one is right; none consolidated the sequence.

This PR keeps every one of those guarantees and removes the round-trips:

1. **`read_file` asks the shell everything in one command.** The probes run inside the same `[ -f ]` stat-not-open guard, so a FIFO or device never reaches `head`/`sed`. Segments are separated by a per-call random sentinel line; "missing" and "not a regular file" travel as in-band markers because a compound command only reports its last exit status. A reply that cannot be parsed (timeout tail, exotic shell) falls back to the previous per-probe sequence, which is preserved as `_read_file_sequential`, so no shell can do worse than today. 4-5 execs to 1.
2. **`write_file` merges its two pre-write probes.** BOM detection (still read from the bytes on disk, never from `pre_content`) and the line-ending / pre-content read ride one command. `_atomic_write` and the `sha256sum` verification are untouched. 4 execs to 3.
3. **Native Python `read_file` on local POSIX environments.** When the env is a `LocalEnvironment`, the read never touches the shell: `os.stat` is the `[ -f ]` guard, the first 1000 bytes drive the same byte-layer binary check, and the page is produced exactly as `sed -n 'a,bp' | cut -b1-N` prints it (byte clamp, newline-terminated) in one streaming pass, then decoded with `errors="replace"` like the transport. `HERMES_NATIVE_FILE_READ=0` turns it off. Windows keeps the shell path because file paths there are held in Git Bash form.

Measured on Linux, `README.md`, 50 lines:

| | per `read_file` | shell calls |
|---|---|---|
| before | 23.6 ms | 4 |
| compound command | 13.9 ms | 1 |
| native local read | 0.15 ms | 0 |

[...]

PR #97770, opened 2026-08-29
fix(search): bound concurrent filename-search CPU work

royalaid:
## Summary
- make `search_files(target="files")` default to bounded discovery order
- preserve exact global modification-time ordering behind `order="modified"`
- resolve installed ripgrep outside inherited PATH and avoid stale negative resolution
- refuse unsafe broad local `find` fallback while keeping narrow, pruned fallback behavior
- serialize same-root filename walks across concurrent file-operation instances with interrupt-safe release
- harden Windows/MSYS paths, multi-root handling, SIGPIPE, error, and ripgrep-version boundaries

## Motivation
When ripgrep is unavailable on inherited PATH, concurrent filename searches can fall through to full GNU `find | sort` walks. Multiple sessions or a single concurrent tool batch can sustain several CPU-heavy scans at once. Modification-time sorting also defeats bounded early output on broad roots.

## Behavior
- default ordering is fast, ignore-aware discovery order
- exact newest-first ordering is explicit through `order="modified"` and requires ripgrep 14+ or GNU `find -printf`
- broad local searches without ripgrep fail before an unbounded fallback walk
- same backend/root scans admit one expensive filename walk at a time; independent roots remain concurrent
- cached `@` completion recency behavior is unchanged

## Bounded zero-match probe
- hidden/ignored zero-match diagnostics retain `--hidden --no-ignore` for useful project-local ignored files, while adding both root and descendant `--glob` exclusions from one shared scan policy covering VCS internals, dependency/vendor trees, virtualenvs, caches, generated/build output, and backup copies
- the real `search_files` regression fixture confirms an ignored `node_modules/package/.hidden/dependency.js` is not traversed or reported while an ignored `.project-local/settings.cfg` remains reported
- an explicit `node_modules/` root is also bounded; hidden descendants are not reported by the fallback diagnostic probe
- Defender evidence motivating this follow-up: 3,739 `rg.exe` scans and 47.6 seconds cumulative Defender scan duration on ignored `node_modules`; no machine-specific local artifact is published

## Verification
- `148 passed, 2 skipped` across filename-search, fallback, macOS exclusion, concurrency, and interrupt suites on native Windows
- `10 passed` across focused `SearchResult`, path-validation, and real Git-Bash fallback classes
- pre-fix RED: the new ignored-tree regression failed because `dependency.js` appeared in the hidden-file warning
- post-fix direct native-Windows focused runs: zero-match/multipath `19 passed`; shared subdirectory-hint policy `56 passed`; filename-search CPU `13 passed`; engine selection `43 passed`; hidden directories `10 passed`; error guard `10 passed`; budget truncation `19 passed`; file-operation edge cases `21 passed`
- Python compilation, staged `git diff --check`, and final commit `git diff HEAD^..HEAD --check` passed
- canonical `scripts/run_tests.sh` was attempted with one worker, but its per-file timeout expired under local host contention before tests completed; that run is not reported as green
- the first post-push Linux CI run found the initial fixture was platform-sensitive because ordinary search could see a non-hidden dependency file; follow-up commit `cc2c991de7` places that file under a hidden descendant, preserving the node_modules-pruning assertion, and its native-Windows focused run passes `19 passed`
- replacement CI run `33285716521` passes the required-check gate, Python tests, Windows/macOS lanes, Python lints, Docker amd64/arm64 builds, nix, OSV, supply-chain, e2e, and repository gates; expected non-applicable jobs are skipped
- unrelated local baseline failures remain outside this change: auto-multiline has 2 failing cases, and file-operation mode/umask tests have 4 failing cases on this Windows environment
- fresh rebase onto `upstream/main@aff5125f8edf5095aef5d3d79bbbb101c95b9413`
- range-diff accounted for all 16 commits: 15 patch-identical, one context-only change from upstream's `file_tools.py` schema refactor

No new dependency is introduced. The focused fallback used the existing Hermes Python 3.11 runtime with pytest packages supplied from a temporary external test target; no test dependency was added to the repository.

### DIFF
--- a/agent/search_policy.py
+++ b/agent/search_policy.py
@@ -1,0 +1,30 @@
+"""Shared directory pruning policy for broad recursive scans.
+
+These names identify version-control internals, dependency trees, generated
+artifacts, caches, and backup copies that are not useful results for broad
+agent-facing discovery. Ordinary search callers may still target an explicit
+path; broad diagnostic probes should apply this policy to recursive walks.
+"""
+
+from __future__ import annotations
+
+
+# Keep this policy conservative and name-based so it works for local and remote
+# shell backends alike. The same set is used by context discovery and search
+# probes; adding a directory here protects every broad recursive consumer.
+SEARCH_PRUNE_DIR_NAMES = frozenset({
+    # Version-control internals.
+    ".git", ".hg", ".svn",
+    # Dependency and vendored trees.
+    "node_modules", "venv", ".venv", "site-packages", "dist-packages",
+    "vendor", "third_party",
+    # Generated/build output.
+    "build", "dist", "target", "out", "coverage",
+    ".next", ".turbo", ".parcel-cache", ".nuxt", ".svelte-kit",
+    # Python and package-manager caches.
+    "__pycache__", ".cache", ".Trash", ".tox", ".nox", ".mypy_cache",
+    ".pytest_cache", ".ruff_cache", ".npm", ".yarn", ".pnpm-store",
+    ".gradle", ".m2", ".nuget",
+    # Backup copies.
+    "backups", "backup", ".backups",
+})
--- a/agent/subdirectory_hints.py
+++ b/agent/subdirectory_hints.py
@@ -21,6 +21,7 @@ from pathlib import Path
 from typing import Dict, Any, Optional, Set
 
 from agent.prompt_builder import _scan_context_content
+from agent.search_policy import SEARCH_PRUNE_DIR_NAMES
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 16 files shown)


kshitijk4poor:
**Triage note (perf/P2 sweep)** — scope check and composition.

- Per-commit review of all 19 commits: every one is on-topic (search ordering / fallback / interrupt / test fixtures) — no dragged-in unrelated batch. Per-file histogram: `tools/file_operations.py` 656 lines, ~1400 in new tests, the remaining 4 source files ≤30 lines each.
- Applies cleanly to `origin/main` 1cb3ab6173; **305 passed, 6 skipped** on macOS.
- **Not reproduced**: the claim is CPU under concurrent `find | sort` fallbacks when rg is off PATH (and Windows Defender scan cost); I could not build a faithful concurrent-fallback fixture within the sweep's time box on macOS, so I am not quoting a number.
- Composition: with #95940 (giant-line containment, same file) the hunks are adjacent only — #95940's diff applies cleanly on top of this branch. With #95160 (probe collapse, same file, canonical for its cluster) it **conflicts** (imports + read path); expected order is #95160 first, then this rebases.

kshitijk4poor:
Merged via #101714 (rebase — all 19 of your commits landed on `main` with authorship preserved; my only changes were import-line conflict resolutions against #95160/#95940 which landed first). Thank you @royalaid — bounded, fail-closed search fallbacks with the shared `search_policy`.

PR #97771, opened 2026-08-29
feat(discord): add profile-scoped LLM DM gate

Dejmacro:
## Summary

Adds `discord.allow_dms` so a profile can keep Discord DMs out of the LLM conversation path without affecting guild-channel admission or slash commands.

This is needed for restricted profiles whose privileged DMs are handled by a separate deterministic broker rather than by the agent. Upstream currently still documents that DMs always get a response (`discord.require_mention` does not apply).

Default remains **enabled** (existing behavior). An explicit profile value wins over the legacy `DISCORD_ALLOW_DMS` env bridge so multiplexed profiles cannot inherit another profile's process-global policy.

## Config

```yaml
discord:
  allow_dms: false
```

## Testing

- `scripts/run_tests.sh tests/plugins/platforms/test_discord_dm_gate.py tests/plugins/platforms/test_discord_gate_isolation.py`
- 36 passed locally (Linux / Python 3.12)
- Contracts covered: default on, config false rejects DMs, legacy env false rejects DMs, profile config precedes env, real `load_gateway_config()` preserves the key, guild messages still admit

## Notes

- Docs point at `config.yaml`; the env var is a migration bridge only.
- No gateway restart / live runtime was used to develop this PR. Cherry-picked from a v0.20.6 local port.

### DIFF
--- a/gateway/config.py
+++ b/gateway/config.py
@@ -1713,6 +1713,8 @@ def load_gateway_config() -> GatewayConfig:
                     bridged["allowed_topics"] = platform_cfg["allowed_topics"]
                 if "free_response_channels" in platform_cfg:
                     bridged["free_response_channels"] = platform_cfg["free_response_channels"]
+                if plat == Platform.DISCORD and "allow_dms" in platform_cfg:
+                    bridged["allow_dms"] = platform_cfg["allow_dms"]
                 if "mention_patterns" in platform_cfg:
                     bridged["mention_patterns"] = platform_cfg["mention_patterns"]
                 if "exclusive_bot_mentions" in platform_cfg:
--- a/plugins/platforms/discord/adapter.py
+++ b/plugins/platforms/discord/adapter.py
@@ -432,6 +432,7 @@ _GATE_ENV_KEYS = (
     "DISCORD_ALLOWED_ROLES",
     "DISCORD_ALLOWED_CHANNELS",
     "DISCORD_IGNORED_CHANNELS",
+    "DISCORD_ALLOW_DMS",
     "DISCORD_NO_THREAD_CHANNELS",
     "DISCORD_FREE_RESPONSE_CHANNELS",
     "DISCORD_MISSED_MESSAGE_BACKFILL_CHANNELS",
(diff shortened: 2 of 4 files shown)


Enough1122:

Reviewed `gateway/config.py`, `plugins/platforms/discord/adapter.py`, new `tests/plugins/platforms/test_discord_dm_gate.py` + docs for PR #97771. Adding profile-scoped `discord.allow_dms` is well-motivated and the tests cover default-on, explicit off (bool and string forms), profile-over-env precedence, guild messages unaffected, and the real gateway loader path. Moving the `is_dm` computation ahead of the role/allowlist checks so the gate fires before any LLM admission is the right placement.

Non-blocking observations:

1. **Group DMs (`GroupChannel`) are treated as DMs.** The gate uses `msg_guild is None` as the DM signal, which is true for both `DMChannel` and Discord group DMs (`GroupChannel`). If `allow_dms: false` is intended to leave group-DM conversations reachable (some restricted profiles want group chats but not 1:1), this silently closes them too. Worth confirming the intended scope and documenting it — the docs only mention "Discord DMs".
2. **An explicit empty/`null` profile value falls through to "enabled".** `if "allow_dms" in extra:` → `raw = extra.get("allow_dms")` → if a user writes `allow_dms:` (YAML null) or an empty string, `raw is None or str(raw).strip() == ""` returns `True`. A typo that yields an empty value silently re-enables DMs on a profile meant to be locked down — failing closed (default to reject on present-but-invalid values) would be safer for a security-gate config. Non-blocking given the explicit-values tests.
3. **`config-first` inverts the documented env-wins rule for this key only.** The docs now say "Security ownership settings such as `discord.allow_dms` are config-first" — coherent and intentional, but a reader following the general "env var wins" contract may be surprised that `DISCORD_ALLOW_DMS=true` cannot re-open a profile that sets `allow_dms: false`. The doc note covers it; just flagging the asymmetry is intentional.
4. **Minor:** `_discord_dms_enabled` is defined right after `_discord_allow_all_users`, which suggests a `_gate_env`-style unified helper could reduce the hand-rolled truthiness parsing (`{"true","1","yes","on"}` vs `{"true","1","yes"}` in `_discord_allow_all_users` — inconsistent accepted-value sets between two adjacent gates).

Solid, well-tested change; nothing blocking.

PR #97772, opened 2026-08-29
feat(camofox): support external noVNC discovery

Dejmacro:
## Summary

Camofox 1.13 moved noVNC metadata from `/health` (`vncPort`) to `/vnc/status` (`novncPort` + path). Hermes still only read the legacy `/health` field, so operators behind a Tailnet reverse proxy never got a usable viewer URL.

This change:

- keeps legacy `/health` `vncPort` discovery
- reads Camofox 1.13 `/vnc/status` when needed
- lets `browser.camofox.vnc_url` override discovered/loopback addresses
- keeps the legacy `CAMOFOX_VNC_URL` env bridge as a fallback
- treats VNC discovery as optional — a missing/failed `/vnc/status` does **not** fail browser health

## Config

```yaml
browser:
  camofox:
    vnc_url: \"https://example.ts.net/vnc.html\"
```

## Testing

- `scripts/run_tests.sh tests/tools/test_browser_camofox_persistence.py`
- 19 passed locally (Linux / Python 3.12)

## Notes

Cherry-picked from a v0.20.6 local port. No live gateway/runtime was modified for this PR.

### DIFF
--- a/tests/tools/test_browser_camofox_persistence.py
+++ b/tests/tools/test_browser_camofox_persistence.py
@@ -269,6 +269,63 @@ class TestVncUrlDiscovery:
             assert check_camofox_available() is True
         assert get_vnc_url() == "http://myhost:6080"
 
+    def test_vnc_url_from_camofox_113_status_endpoint(self, monkeypatch):
+        monkeypatch.setenv("CAMOFOX_URL", "http://myhost:9377")
+        health_resp = _mock_response(json_data={"ok": True})
+        status_resp = _mock_response(json_data={
+            "running": True,
+            "novncPort": 6080,
+            "path": "/vnc.html",
+        })
+        with patch(
+            "tools.browser_camofox.requests.get",
+            side_effect=[health_resp, status_resp],
+        ) as mock_get:
+            assert check_camofox_available() is True
+        assert mock_get.call_args_list[1].args[0] == "http://myhost:9377/vnc/status"
+        assert get_vnc_url() == "http://myhost:6080/vnc.html"
+
+    def test_config_external_vnc_url_overrides_discovery_and_legacy_env(
+        self, monkeypatch
+    ):
+        monkeypatch.setenv("CAMOFOX_URL", "http://127.0.0.1:9377")
+        monkeypatch.setenv("CAMOFOX_VNC_URL", "http://legacy.example:6080/vnc.html")
+        config = {"browser": {"camofox": {
+            "vnc_url": "https://gx10.example.ts.net/vnc.html",
+        }}}
+        health_resp = _mock_response(json_data={"ok": True})
+        with (
+            patch("tools.browser_camofox.load_config", return_value=config),
+            patch(
+                "tools.browser_camofox.requests.get", return_value=health_resp
+            ) as mock_get,
+        ):
+            assert check_camofox_available() is True
+        assert mock_get.call_count == 1
+        assert get_vnc_url() == "https://gx10.example.ts.net/vnc.html"
+
+    def test_legacy_external_vnc_url_bridge(self, monkeypatch):
+        monkeypatch.setenv("CAMOFOX_URL", "http://127.0.0.1:9377")
+        monkeypatch.setenv(
+            "CAMOFOX_VNC_URL", "https://gx10.example.ts.net/vnc.html"
+        )
+        health_resp = _mock_response(json_data={"ok": True})
+        with patch("tools.browser_camofox.requests.get", return_value=health_resp):
+            assert check_camofox_available() is True
+        assert get_vnc_url() == "https://gx10.example.ts.net/vnc.html"
+
+    def test_optional_vnc_status_failure_does_not_fail_browser_health(
+        self, monkeypatch
+    ):
+        monkeypatch.setenv("CAMOFOX_URL", "http://myhost:9377")
+        health_resp = _mock_response(json_data={"ok": True})
+        with patch(
+            "tools.browser_camofox.requests.get",
+            side_effect=[health_resp, RuntimeError("optional endpoint unavailable")],
+        ):
+            assert check_camofox_available() is True
+        assert get_vnc_url() is None
+
 
     def test_navigate_includes_vnc_hint(self, tmp_path, monkeypatch):
         monkeypatch.setenv("HERMES_HOME", str(tmp_path))
--- a/tools/browser_camofox.py
+++ b/tools/browser_camofox.py
@@ -149,29 +149,78 @@ def is_camofox_mode() -> bool:
     return bool(get_camofox_url())
 
 
+def _validated_vnc_url(raw: Any) -> Optional[str]:
+    """Return a safe absolute HTTP(S) VNC URL, or ``None``."""
+    value = str(raw or "").strip()
+    if not value:
+        return None
+    try:
+        parsed = urlsplit(value)
+    except ValueError:
+        return None
+    if parsed.scheme not in {"http", "https"} or not parsed.hostname:
+        return None
+    return value
+
+
+def _discovered_vnc_url(server_url: str, port: Any, path: str = "") -> Optional[str]:
+    """Build an HTTP(S) VNC URL from validated server metadata."""
+    if not isinstance(port, int) or isinstance(port, bool) or not 1 <= port <= 65535:
+        return None
+    parsed = urlsplit(server_url)
+    host = parsed.hostname or "localhost"
+    if ":" in host and not host.startswith("["):
+        host = f"[{host}]"
+    scheme = parsed.scheme if parsed.scheme in {"http", "https"} else "http"
+    safe_path = path if path.startswith("/") and not path.startswith("//") else ""
+    return f"{scheme}://{host}:{port}{safe_path}"
+
+
 def check_camofox_available() -> bool:
-    """Verify the Camofox server is reachable."""
+    """Verify Camofox health and opportunistically cache its noVNC URL."""
     global _vnc_url, _vnc_url_checked
     url = get_camofox_url()
     if not url:
         return False
     try:
         resp = requests.get(f"{url}/health", timeout=5)
-        if resp.status_code == 200 and not _vnc_url_checked:
-            try:
-                data = resp.json()
-                vnc_port = data.get("vncPort")
-                if isinstance(vnc_port, int) and 1 <= vnc_port <= 65535:
-                    from urllib.parse import urlparse
-                    parsed = urlparse(url)
-                    host = parsed.hostname or "localhost"
-                    _vnc_url = f"http://{host}:{vnc_port}"
-            except (ValueError, KeyError):
-                pass
-            _vnc_url_checked = True
-        return resp.status_code == 200
     except Exception:
         return False
+    if resp.status_code != 200:
+        return False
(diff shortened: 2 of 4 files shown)


Enough1122:

Reviewed `tools/browser_camofox.py` + tests + docs for PR #97772. The refactor is clean: health check and VNC discovery are now decoupled, the legacy `/health` `vncPort` path is preserved, Camofox 1.13's `/vnc/status` is consulted as an optional fallback, and both `browser.camofox.vnc_url` config and the legacy `CAMOFOX_VNC_URL` env bridge are honored with explicit config precedence. `_validated_vnc_url` and `_discovered_vnc_url` are good hardening (scheme/hostname/port bounds, IPv6 brackets, path validation), and tests cover config-over-env, legacy env, and the optional-endpoint-failure case.

Non-blocking observations:

1. **`_get_camofox_config()` / `get_secret()` now run outside the exception guard.** The old function wrapped its whole body in `try/except Exception: return False`; the refactor only guards the `requests.get` and the two JSON-parsing blocks. `_get_camofox_config()` and `get_secret("CAMOFOX_VNC_URL", "")` run in the `if not _vnc_url_checked` block with no guard, so a config-loading or secret-file error would now propagate instead of returning `False`. Callers presumably treat this as a pure boolean probe — consider wrapping the whole discovery block (or at least `_get_camofox_config`) so health probing can't throw.
2. **The `_vnc_url_checked` cache is never invalidated.** Once set, the discovered URL is cached for process lifetime. If the Camofox server restarts on a different port or the operator changes `browser.camofox.vnc_url` at runtime, `get_vnc_url()` keeps returning the stale value. This matches the prior behavior (cache existed before), so it's not a regression — just noting the explicit `vnc_url` config would still require a restart to take effect.
3. **Path is not URL-encoded and may carry no query.** `_discovered_vnc_url` accepts any `path` starting with `/` (and not `//`), so a Camofox-returned path containing spaces or special characters would be embedded raw. The `/vnc/status` default is `/vnc.html`, so in practice this is benign; only flagging for the proxy-with-auth-query case where an unencoded path could 404.
4. **Minor:** `data.get("vncPort")` inside the `resp.json()` block — if a proxy ever returned a non-object JSON, `resp.json()` could yield a non-dict and `.get` would raise `AttributeError`, which isn't in the caught tuple `(TypeError, ValueError, KeyError)`. A broader `except Exception` (as used on the `/vnc/status` branch) would be more uniform.

Well-structured and well-tested; nothing blocking.

DavidMetcalfe:
@Dejmacro — the direction is right, and I can confirm the contract from a running server rather than from source alone.

I stood camofox-browser up locally (master `79d425be`). `GET /health` returns a fixed field set with no VNC key; with `ENABLE_VNC=1`, `GET /vnc/status` returns `{"enabled":true,"running":false,"watcherRunning":false,"vncPort":5900,"novncPort":6080,"path":"/vnc.html"}`, and 404s when the plugin is off. Building the link from `novncPort` + `path` is right — `vncPort` is the x11vnc port and isn't browsable.

**The branch no longer applies cleanly, and the obvious resolution reverts a fix that landed after you opened it.**

`git merge-tree --write-tree --name-only origin/main <head>` flags one conflicted file, `tools/browser_camofox.py` (the other three auto-merge). Cause: #108935 ([`4b8c01f691`](https://github.com/NousResearch/hermes-agent/commit/4b8c01f691), merged 2026-09-12) rewrote the same two functions so the discovered URL is memoized per Camofox server in `_vnc_url_by_camofox_url` when `get_hermes_home_override() is not None` — a multiplexed gateway serves several profiles from one process, each pointing `CAMOFOX_URL` at its own server.

Your base predates that, and your side of the hunk writes only the one-shot global. Take your side and `get_vnc_url()`'s multiplex branch (your PR doesn't touch it, so it survives the merge) finds `_vnc_url_by_camofox_url` permanently empty and returns `None` for every routed profile: #108935's fix is gutted while the dict lingers as dead code.

**A resolution that keeps both** — your discovery logic lifted into a helper, main's dual-memo dispatch kept, the superseded `_vnc_url_from_health` removed:

<details><summary>diff against origin/main</summary>

```diff
diff --git a/tools/browser_camofox.py b/tools/browser_camofox.py
index 8618a595c6..0258de0182 100644
--- a/tools/browser_camofox.py
+++ b/tools/browser_camofox.py
@@ -110,18 +110,74 @@ def is_camofox_mode() -> bool:
     return bool(get_camofox_url())
 
 
-def _vnc_url_from_health(url: str, resp: Any) -> Optional[str]:
+def _validated_vnc_url(raw: Any) -> Optional[str]:
+    """Return a safe absolute HTTP(S) VNC URL, or ``None``."""
+    value = str(raw or "").strip()
+    if not value:
+        return None
     try:
-        vnc_port = resp.json().get("vncPort")
-        if isinstance(vnc_port, int) and 1 <= vnc_port <= 65535:
-            return f"http://{urlsplit(url).hostname or 'localhost'}:{vnc_port}"
-    except (ValueError, KeyError):
-        pass
-    return None
+        parsed = urlsplit(value)
+    except ValueError:
+        return None
+    if parsed.scheme not in {"http", "https"} or not parsed.hostname:
+        return None
+    return value
+
+
+def _discovered_vnc_url(server_url: str, port: Any, path: str = "") -> Optional[str]:
+    """Build an HTTP(S) VNC URL from validated server metadata."""
+    if not isinstance(port, int) or isinstance(port, bool) or not 1 <= port <= 65535:
+        return None
+    parsed = urlsplit(server_url)
+    host = parsed.hostname or "localhost"
+    if ":" in host and not host.startswith("["):
+        host = f"[{host}]"
+    scheme = parsed.scheme if parsed.scheme in {"http", "https"} else "http"
+    safe_path = path if path.startswith("/") and not path.startswith("//") else ""
+    return f"{scheme}://{host}:{port}{safe_path}"
+
+
+def _discover_vnc_url(url: str, resp: Any) -> Optional[str]:
+    """Resolve the noVNC URL for ``url`` from the ``/health`` response ``resp``.
+
+    Precedence: ``browser.camofox.vnc_url`` config override, then the legacy
+    ``CAMOFOX_VNC_URL`` env bridge, then the legacy ``vncPort`` field on
+    ``/health``, then Camofox 1.13's optional ``/vnc/status`` endpoint.
+    """
+    camofox_cfg = _get_camofox_config()
+    vnc_url = (
+        _validated_vnc_url(camofox_cfg.get("vnc_url"))
+        or _validated_vnc_url(get_secret("CAMOFOX_VNC_URL", ""))
+    )
+    if vnc_url is None:
+        try:
+            data = resp.json()
+            vnc_url = _discovered_vnc_url(url, data.get("vncPort"))
+        except (TypeError, ValueError, KeyError):
+            vnc_url = None
+
+    # Camofox 1.13 moved noVNC metadata from /health to /vnc/status.
+    # This endpoint is optional: a failure must not make browser health fail.
+    if vnc_url is None:
+        try:
+            status_resp = requests.get(f"{url}/vnc/status", timeout=5)
+            if status_resp.status_code == 200:
+                status = status_resp.json()
+                path = status.get("path")
+                if not isinstance(path, str):
+                    path = "/vnc.html"
+                vnc_url = _discovered_vnc_url(
+                    url,
+                    status.get("novncPort"),
+                    path,
+                )
+        except Exception:
+            pass
+    return vnc_url
 
 
 def check_camofox_available() -> bool:
-    """Verify the Camofox server is reachable (and cache its VNC URL once)."""
+    """Verify Camofox health and opportunistically cache its noVNC URL."""
     global _vnc_url, _vnc_url_checked
     url = get_camofox_url()
     if not url:
@@ -130,14 +186,19 @@ def check_camofox_available() -> bool:
         resp = requests.get(f"{url}/health", timeout=5)
     except Exception:
         return False
-    if resp.status_code == 200:
-        if get_hermes_home_override() is not None:
-            if url not in _vnc_url_by_camofox_url:
-                _vnc_url_by_camofox_url[url] = _vnc_url_from_health(url, resp)
-        elif not _vnc_url_checked:
-            _vnc_url = _vnc_url_from_health(url, resp) or _vnc_url
-            _vnc_url_checked = True
-    return resp.status_code == 200
+    if resp.status_code != 200:
+        return False
+
+    if get_hermes_home_override() is not None:
+        # Multiplexed gateway: several profiles share this process, each points
+        # CAMOFOX_URL at its own server — memo per server URL, never the
+        # one-shot global slot (see PR #108935).
+        if url not in _vnc_url_by_camofox_url:
+            _vnc_url_by_camofox_url[url] = _discover_vnc_url(url, resp)
+    elif not _vnc_url_checked:
+        _vnc_url = _discover_vnc_url(url, resp) or _vnc_url
+        _vnc_url_checked = True
+    return True
 
 
 def get_vnc_url() -> Optional[str]:
```
</details>

Verification with that resolution, using the repo's file-granular runner:

- `tests/tools/test_browser_camofox_persistence.py` → 19 passed
- `-k vnc` → 6 passed
- `tests/tools/test_multiplex_tool_cache_scope.py` (the #108935 regression file) → 4 passed

The tests do exercise the missing behaviour: against main's `tools/browser_camofox.py` with your test file, `-k vnc` is red — 3 failed, 3 passed — with `test_vnc_url_from_camofox_113_status_endpoint` asserting on a second request that never happens and the two override tests observing `None`.

Two smaller points:

- Nothing in the PR sets a hermes-home override, so multiplex + `/vnc/status` is untested — on `main` as well. One test with the override active and a `/vnc/status` response would pin the path this conflict lives in.
- `CAMOFOX_VNC_URL` doesn't exist on `main` today. There is precedent for the pattern — `CAMOFOX_URL` is read from the environment — but the contribution rubric asks behavioural settings to live in `config.yaml` rather than env vars, and `browser.camofox.vnc_url` already covers the reverse-proxy case. Whether to keep the env bridge alongside the config key looks like a maintainer call.

The rebase is the only thing standing between this and a reviewable diff — happy to re-run the verification against your rebased head.

DavidMetcalfe:
@Dejmacro — flagging that I've salvaged this PR, with your commit and credit preserved.

Your read of the contract is right, and I verified it against a live camofox-browser (master `79d425be`): `/health` carries no VNC field, the plugin serves that metadata on `GET /vnc/status`, and the browsable address is `novncPort` + `path` rather than the raw x11vnc `vncPort`.

What blocked it here was the base. It predates #108935 ([`4b8c01f691`](https://github.com/NousResearch/hermes-agent/commit/4b8c01f691)), which rewrote `check_camofox_available()` / `get_vnc_url()` so the discovered URL is memoized per Camofox server while a home override is active — a multiplexed gateway serves several profiles from one process, each with its own `CAMOFOX_URL`. Taking your side of that hunk drops the guard: `_vnc_url_by_camofox_url` stays empty and `get_vnc_url()` returns `None` for every routed profile.

Since the branch has been quiet since August, I've opened #118623: your commit cherry-picked onto current `main` (it still lists you as the author), plus a follow-up commit that drops the `CAMOFOX_VNC_URL` env bridge in favour of the `browser.camofox.vnc_url` config key, declares that key in `config_defaults.py`, and adds a regression test for the multiplex path this conflict lives in. The defect itself is filed as #118619.

Thanks for the original work — the `/vnc/status` approach was the part that made this fixable.

This PR stays open; closing it is your call or a maintainer's.

PR #97773, opened 2026-08-29
fix(dashboard): forward startup query to chat PTY

Dejmacro:
## Summary

The TUI already honors `HERMES_TUI_QUERY`, but the web dashboard PTY path never forwarded the authenticated WebSocket `query` parameter into the child env. Dashboard deep-links therefore could not seed the chat the way the CLI/TUI can.

This change:

- reads `query` from the PTY WebSocket (trimmed, capped at 4000 chars)
- forwards it as `HERMES_TUI_QUERY` for that PTY only
- pops any inherited `HERMES_TUI_QUERY` from the dashboard service env so a stale service-level value cannot leak into every chat

## Testing

- `scripts/run_tests.sh tests/hermes_cli/test_web_server.py -k 'startup_query or query_parameter'`
- 2 passed locally (Linux / Python 3.12)

## Notes

Internal env bridge only; no new user-facing `.env` setting. Cherry-picked from a v0.20.6 local port.

### DIFF
--- a/hermes_cli/web_server.py
+++ b/hermes_cli/web_server.py
@@ -16598,6 +16598,7 @@ def _resolve_chat_argv(
     sidecar_url: Optional[str] = None,
     profile: Optional[str] = None,
     active_session_file: Optional[str] = None,
+    query: Optional[str] = None,
 ) -> tuple[list[str], Optional[str], Optional[dict]]:
     """Resolve the argv + cwd + env for the chat PTY.
 
@@ -16682,6 +16683,11 @@ def _resolve_chat_argv(
     # the dashboard PTY path.
     env.setdefault("HERMES_TUI_DISABLE_MOUSE", "1")
     env.setdefault("HERMES_TUI_INLINE", "1")
+    # Never inherit a stale startup query from the dashboard service. Only
+    # the authenticated WebSocket request for this PTY may seed one.
+    env.pop("HERMES_TUI_QUERY", None)
+    if query:
+        env["HERMES_TUI_QUERY"] = query
     # The dashboard terminal is xterm.js, which always renders 24-bit RGB.
     # But chalk inside the TUI child decides its color depth from the
     # SERVER process env — and hosted/cloud deploys run the dashboard under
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed `hermes_cli/web_server.py` + tests for PR #97773. Forwarding the authenticated WebSocket `query` param into the dashboard chat PTY's env is a clean, well-scoped fix: `_resolve_chat_argv` now pops any inherited `HERMES_TUI_QUERY` before setting the per-PTY one (preventing a stale service-level value from leaking into unrelated PTYs), the async resolver threads the param through, and the WS handler trims and caps it at 4000 chars. Tests cover the sync resolver env injection and the WS→resolver hand-off (including whitespace trimming).

Non-blocking observations:

1. **`_resolve_chat_argv` is dashboard-only, so the unconditional `env.pop` is safe.** Confirmed `_resolve_chat_argv`/`_resolve_chat_argv_async` are used only by the web-server PTY path (the CLI/TUI uses `_make_tui_argv` directly), so always popping `HERMES_TUI_QUERY` can't strip a user-set seed from a non-dashboard path. Good.
2. **Cap is applied in `pty_ws`, not at the resolver.** `query = (...)[:4000]` in `pty_ws`, while `_resolve_chat_argv` sets `env["HERMES_TUI_QUERY"] = query` unguarded. Any future caller of `_resolve_chat_argv` that passes an unbounded `query` would bypass the cap. Since the param already has a `:4000` ceiling at the single call site today this is fine — just worth a comment or a `[:4000]` at the env-set site so the invariant is enforced where it matters.
3. **Empty/whitespace query is dropped at the WS layer but `pop` still runs in the resolver.** `("" or None)` → no `query` kwarg → `env.pop("HERMES_TUI_QUERY")` still executes, which is exactly the intended "no stale seed" behavior. Consistent.
4. **Minor:** `_resolve_chat_argv_async` only forwards `query` when truthy; a caller passing an intentionally empty string to *clear* an inherited query would get the pop path via the sync resolver anyway, so no behavioral gap.

Correct, minimal, and well-tested; nothing blocking.

PR #96527, opened 2026-08-27
feat(discord): opt-in voice auto-join with safety gates

DavidMetcalfe:
## Summary

Partially addresses #33683 (Phase 2 — opt-in auto-join policy).

Adds a disabled-by-default Discord voice auto-join mode, gated on every side:

```yaml
discord:
  voice_auto_join:
    enabled: false                  # master switch — must be explicitly true
    channel_ids: []                 # REQUIRED non-empty: VC IDs the bot may auto-join
    user_ids: []                    # REQUIRED non-empty: users whose VC entry triggers auto-join
    join_mode: "user_prompt"        # "user_prompt" (approval reaction required) | "automatic"
    require_text_opt_in: true       # triggering user must be able to read the bound text channel
```

## Safety gates

- **Both allowlists required.** Activation refuses to run unless `channel_ids` *and* `user_ids` are non-empty. This is deliberate: without it, a user who enables auto-join without an explicit allowlist would hit the existing allow-all fallthrough and transcribe every speaker in the channel.
- **Fail-closed approval prompt.** In `user_prompt` mode (default), the bot posts an approval message with ✅/❌ reactions and only joins on ✅ (60 s window, scoped to the triggering member and that message). If no suitable text channel exists or the prompt cannot be sent, the join is **aborted**, never silently executed — failure to prompt is failure to join.
- **Loop prevention.** The bot-self guard in `on_voice_state_update` runs first and unconditionally returns (covers bot-initiated joins *and* admin-initiated bot moves, since the fired event's member is the bot itself). Auto-join/leave evaluation is debounced per (guild, member, channel).
- **Announced joins.** Every successful auto-join posts an announcement in the bound text channel ("🤖 Hermes auto-joined \<#…\> — transcribing for allowlisted users. Use `/voice leave` to disconnect.").
- **Leave conditions.** Existing idle timeout is untouched; auto-join adds leave-when-no-authorized-users, evaluated only when the last left channel is the one the bot is actually connected to (a user leaving an unrelated VC in the same guild does not disconnect the bot).
- **Wrong-channel protection.** The channel allowlist is checked by ID pre-join; `join_voice_channel`'s existing move-to behavior is unchanged.

## Design

- `plugins/platforms/discord/voice_policy.py` (new) — pure-logic, zero discord.py imports: frozen `AutoJoinPolicy` dataclass + `VoicePolicyEvaluator` (`activation_error`, `should_join`). Same sibling-module pattern as `voice_mixer.py`; trivially unit-testable.
[...]

PR #67225, opened 2026-07-19
fix(discord): make voice auto-join profile-safe and transactional

marcelomediagroup:
## What does this PR do?

Ships the fixed-point Discord voice auto-join candidate from internal OE-98. It preserves the follow-mode behavior introduced in #53747 while hardening the full lifecycle: profile and guild ownership, target validation, join token gating, overlapping joins, retry cancellation, leave/disconnect races, transactional teardown, rollback, and fail-closed outcomes.

This PR explicitly supersedes #53747. That predecessor is now closed; this successor is the single current review surface and carries the reviewed implementation plus the complete hardening and regression chain at exact head `f29524c82`.

## Related Issue

- Internal tracker: OE-98
- Supersedes #53747
- Duplicate-scope audit: #59885 and #62499 overlap voice auto-join behavior. They remain open and unchanged; this PR is specifically the hardened successor to our #53747 implementation.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [x] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Adds configurable Discord voice auto-join follow mode and documents its configuration.
- Scopes callbacks, mode state, deduplication, and session ownership by adapter profile and guild.
- Validates allowed users, text/voice guild relationships, and target channels before joining.
- Makes joins token-gated and fail-closed, including legacy-looking spoof errors and incompatible call signatures.
- Serializes join, move, leave, teardown, rollback, and retry cleanup under the voice lock.
- Adds event-controlled regression coverage for overlap, disconnect, mid-move rollback, cleanup, and retry races.

## How to Test

1. `scripts/run_tests.sh tests/gateway/test_voice_command.py -q` — 256/256 passed at `f29524c82`.
2. `scripts/run_tests.sh tests/gateway/test_discord_missed_message_backfill.py -q` — 44/44 passed at the same head.
3. The complete Discord slice passed 594 tests.
4. The config/voice integration slice passed 384 tests.
5. Ruff, bytecode compilation, the Windows-footgun check, diff checks, and ancestry checks passed.
6. Independent Spec and Standards reviews approved the exact rebased head with zero findings.

No live Discord server was mutated during verification; lifecycle behavior is exercised through instrumented/event-controlled integration tests.

[...]

PR #62499, opened 2026-07-11
feat(discord): production-grade voice channel with configurable VAD and auto-join

jerryopncld:
## Summary

Add mode-aware NaCl decryption, RMS-based voice activity detection (VAD), auto-join/auto-disconnect, and voice_states-based user lookup that works without the privileged Members intent.

### Mode-aware NaCl decryption ()
- Support all 4 discord.py encryption modes: `aead_xchacha20_poly1305_rtpsize`, `xsalsa20_poly1305_lite`, `xsalsa20_poly1305_suffix`, `xsalsa20_poly1305`
- Using the wrong cipher/nonce format silently fails decryption — the mode is now read from the connection and handled per-mode
- Mid-session key rotation: Discord can send `VOICE_SERVER_UPDATE` which triggers a reconnect with a new secret key. The receiver now refreshes key/ssrc/mode on every packet

### RMS-based voice activity detection ()
- Energy-based VAD filters comfort-noise/silence frames using an RMS threshold (~-40dBFS)
- Prevents tens of seconds of silence from inflating recordings and breaking the silence-end-of-utterance trigger
- Only buffers frames with voice energy; silence frames during active utterance are buffered for natural transitions

### Auto-join / auto-disconnect (, `run.py`)
- Bot auto-joins when an allowed user enters a voice channel (configurable: `discord.voice.auto_join`)
- Bot auto-disconnects when the last human leaves (configurable: `discord.voice.auto_disconnect`)
- Optional specific voice channel filter (`discord.voice.channel_id`) — bot only auto-joins this channel
- Gateway wires voice callbacks (input handler, TTS, disconnect handler) on auto-join via `_wire_voice_auto_join`

### voice_states-based user lookup ()
- `get_user_voice_channel` now uses `guild._voice_states` (populated from `VOICE_STATE_UPDATE` events) which only requires `intents.voice_states = True`
- Falls back to member cache/fetch only if voice_states doesn't have the user
- Works without the privileged Server Members intent

### VoiceMixer ()
- Properly subclasses `discord.AudioSource` with graceful fallback when discord.py is not installed

### STT tuning (`transcription_tools.py`)
- `beam_size`, `vad_filter`, `cpu_threads`, and `initial_prompt` are now configurable via `stt.local.*` in config.yaml

## Configuration

All voice parameters are configurable via `discord.voice.*` in config.yaml:

```yaml
discord:
  voice:
    auto_join: false          # auto-join when allowed user enters voice channel
[...]

PR #59885, opened 2026-07-06
feat: auto-join Discord voice channel when allowed user joins

mari-meiji:
## Problem

Discord voice channels currently require a user to manually type `/voice join` every single time they want the bot to join, even in servers where the bot's owner always wants it in voice alongside them. This is a small but repetitive piece of friction on every voice session.

## Solution

Adds an **opt-in** `discord.voice.auto_join_on_user_join` config flag (default `false`). When enabled, `on_voice_state_update` calls the adapter's existing `join_voice_channel()` — the exact same method `/voice join` calls, including its per-guild lock, `VoiceReceiver` startup, and mixer install — as soon as a qualifying user enters a voice channel the bot isn't already connected to.

### Config keys (config.yaml, not .env — per AGENTS.md conventions)

```yaml
discord:
  voice:
    auto_join_on_user_join: false   # opt-in master switch (default off)
    auto_join_users: []             # optional trigger allowlist; empty = any allowed user
```

- `auto_join_on_user_join` — master switch, default `false`.
- `auto_join_users` — optional list of Discord user IDs/usernames. Empty (default) means any user who already passes the existing `DISCORD_ALLOWED_USERS` / `DISCORD_ALLOWED_ROLES` gate can trigger auto-join. When non-empty, only listed users trigger it (they must still pass the existing allowlist too).

### Backward compatibility

- Default is `false` — zero behavior change for existing users unless they explicitly opt in.
- Purely additive: `/voice join` and `/voice leave` slash commands are untouched and keep working identically.
- Never triggers for bot members joining voice.
- No-ops if the bot is already connected to the channel the user joined (avoids redundant `join_voice_channel()` calls).
- Reuses `join_voice_channel()`'s existing per-guild `asyncio.Lock` — no new locking primitive introduced.

## Files changed

- `hermes_cli/config.py` — new `discord.voice.{auto_join_on_user_join,auto_join_users}` keys in `DEFAULT_CONFIG`.
- `plugins/platforms/discord/adapter.py` — `_load_voice_auto_join_config()` (mirrors the existing `_load_voice_fx_config()` pattern), plus the additive auto-join branch inside `connect()`'s `on_voice_state_update` handler.
- `tests/gateway/test_discord_voice_auto_join.py` — new test module.
- `website/docs/user-guide/features/voice-mode.md` — new "Auto-Join" subsection documenting the config keys, table-driven to match the rest of the page.

## Test coverage

[...]

PR #97774, opened 2026-08-29
feat(discord): add profile-scoped voice auto-join

Dejmacro:
## Summary

Adds opt-in, **disabled-by-default** Discord voice auto-join that fires only after an admitted **live guild** message, joining the sender's current voice channel and enabling spoken replies for that text chat.

This is the smaller engagement-on-message design already running on our v0.20.6 port. It is **not** a duplicate of the `on_voice_state_update` allowlist PRs:

- #96527 opt-in voice auto-join with channel/user allowlists + approval prompt
- #59885, #67225, #62499 related voice auto-join variants

Those cover join-on-VC-entry. This PR covers join-on-admitted-text-engagement, which is what `/voice join` already does, just triggered from a live guild message when explicitly enabled.

## Safety

- Default **off**
- Recovered/history backfill never auto-joins
- Profile `discord.voice_auto_join` wins over the legacy `DISCORD_VOICE_AUTO_JOIN` env bridge
- `_apply_yaml_config` skips the env write under profile-scoped multiplex loads
- Real `load_gateway_config()` probe so `PlatformConfig.extra` actually keeps the key

## Config

```yaml
discord:
  voice_auto_join: true
```

## Testing

- `scripts/run_tests.sh tests/gateway/test_discord_voice_auto_join.py tests/plugins/platforms/test_discord_gate_isolation.py`
- 28 passed locally (Linux / Python 3.12)

Happy to close this as duplicate if maintainers prefer #96527 as the only auto-join surface.

### DIFF
--- a/gateway/config.py
+++ b/gateway/config.py
@@ -1713,6 +1713,8 @@ def load_gateway_config() -> GatewayConfig:
                     bridged["allowed_topics"] = platform_cfg["allowed_topics"]
                 if "free_response_channels" in platform_cfg:
                     bridged["free_response_channels"] = platform_cfg["free_response_channels"]
+                if plat == Platform.DISCORD and "voice_auto_join" in platform_cfg:
+                    bridged["voice_auto_join"] = platform_cfg["voice_auto_join"]
                 if "mention_patterns" in platform_cfg:
                     bridged["mention_patterns"] = platform_cfg["mention_patterns"]
                 if "exclusive_bot_mentions" in platform_cfg:
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -13449,6 +13449,14 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                 # transcription is forwarded without requiring /voice join.
                 if hasattr(adapter, "_voice_input_callback"):
                     adapter._voice_input_callback = self._handle_voice_channel_input
+                if hasattr(adapter, "_on_voice_disconnect"):
+                    adapter._on_voice_disconnect = self._handle_voice_timeout_cleanup
+                if hasattr(adapter, "_on_voice_joined"):
+                    adapter._on_voice_joined = self._handle_voice_joined
+                if hasattr(adapter, "_voice_mode_getter"):
+                    adapter._voice_mode_getter = lambda chat_id: self._voice_mode.get(
+                        self._voice_key(Platform.DISCORD, str(chat_id)), "off"
+                    )
                 connected_count += 1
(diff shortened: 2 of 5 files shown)


Enough1122:

Reviewed `plugins/platforms/discord/adapter.py`, `gateway/run.py`, `gateway/config.py`, new `tests/gateway/test_discord_voice_auto_join.py` + docs for PR #97774. The opt-in, disabled-by-default voice auto-join is well-contained: it fires only on a non-recovered live guild message when `voice_auto_join` is enabled for the profile, joins the sender's current VC, re-binds the text channel/source, and notifies the runner to enable persisted voice mode + auto-TTS. Tests cover config seeding, the real loader, profile-over-env precedence, env-snapshot support, bind+notify, disabled-no-probe, and the runner callback.

Non-blocking observations:

1. **No short-circuit when the bot is already in that guild's voice channel.** `maybe_auto_join_voice` runs `get_user_voice_channel` (a Discord API member lookup) on every admitted live guild message even when the bot is already in the same guild VC with the same text channel bound. On a busy guild that's one API call per message and a repeated `join_voice_channel`/`_reset_voice_timeout` per message. A cheap early return (e.g. `if self._voice_text_channels.get(guild_id) == int(text_channel_id) and self.is_in_voice_channel(guild_id): return True`) would cut the per-message load with no behavior change.
2. **The auto-join path is gated only by "non-recovered guild message", not by the author.** Any admitted member's message (subject to whatever admission rules apply before `_handle_message`) can pull the bot into their VC. That's the stated design ("smaller engagement-on-message"), and it's opt-in per profile, but it means a single user can repeatedly relocate the bot (e.g. join/leave VCs and message) — consider a rate limit / cooldown or restricting to users who can already control the bot (allowlist), especially on public servers. Non-blocking.
3. **Every auto-join can persist voice mode.** `_handle_voice_joined` writes `_voice_mode` via `_save_voice_modes()` when the mode isn't already `"all"`. If a user leaves and rejoins VCs repeatedly, this is a disk write per transition. Minor, but a guard that skips the save when the mode is already `"all"` is already present (`if ... != "all"`), so it's only the first transition per chat — fine.
4. **`_apply_yaml_config` bridges to `DISCORD_VOICE_AUTO_JOIN` env even when `_skip_env_bridge` is unset but the env is already populated.** The bridge sets `os.environ["DISCORD_VOICE_AUTO_JOIN"] = str(voice_auto_join).lower()` only when the env isn't already set — good, preserves an explicit env override. The `_skip_env_bridge` guard adds a second exit. Consistent with the allow_dms PR's approach.

The core flow is correct and well-tested; nothing blocking.

PR #97775, opened 2026-08-29
fix(gateway): honor --global in reasoning picker

ricardosalta:
## What does this PR do?

`/reasoning --global` (including iOS-style `/reasoning —global`) currently
returns an unknown-empty-argument error. Treat the parsed empty value as a
picker request and carry `persist_global` into its selection callback, so
choosing an effort saves it to config and clears the conversation override.

The delayed callback captures and re-enters the originating multiplex profile
scope, matching `/model` and using the existing profile-aware config writer.
This prevents a named profile's picker selection from modifying the default
profile after the original command scope has ended.

## Type of Change

- [x] Bug fix

## Changes Made

- `gateway/slash_commands_model.py`: use the parsed value for dispatch; preserve global intent and callback profile scope.
- `tests/gateway/test_choice_picker.py`: two invariant regressions, parameterized for ASCII/em-dash flags, exercising real YAML persistence and delayed profile isolation.

## How to Test

Send `/reasoning --global` or `/reasoning —global`, then select an effort.
Verify `agent.reasoning_effort` is saved in the originating profile and the
session override is cleared. Repeat with a named multiplex profile.

Validated on Linux using `scripts/run_tests.sh -j 2 tests/gateway/test_choice_picker.py tests/gateway/test_reasoning_command.py tests/gateway/test_reasoning_config_per_model.py -q` (21 passed). Both flag spellings failed on pinned upstream before the fix; the delayed profile test independently failed after correcting only flag dispatch/persistence. Ruff 0.15.10, Windows-footgun checks and plugin compatibility-pointer checks passed. The full repository suite was not run.


Enough1122:

Reviewed `gateway/slash_commands.py` + tests for PR #97775. The root fix is the `if not raw_args:` → `if not args:` change: `args` is empty after `_parse_reasoning_command_args` strips a bare `--global` (both ASCII and em-dash forms), so a flag-only global invocation now correctly opens the picker instead of falling into the typed-arg path. The multiplex-scope plumbing is also right: `_command_profile_home` is resolved once at command time (so a *delayed* picker tap writes to the originating profile, not wherever the process scope is at tap time), and the picker callback now forwards `persist_global` — a real gap, since the picker path previously never persisted globally. Tests cover both flag spellings, global persistence, and the named-profile isolation (including a delayed tap after `_profile_runtime_scope` exits).

Non-blocking observations:

1. **Override-clearing happens on `session_key`, which is profile-scoped, but the write target is `config_home`.** In the multiplex delayed-tap case, `_apply_reasoning_selection` clears `_session_reasoning_overrides[session_key]` and calls `_evict_cached_agent(session_key)` while writing config to the originating profile home. If the tap fires after the multiplex has switched the active profile, `session_key` may belong to the *other* profile — the config write lands correctly in the originating profile, but the override/cache eviction may target the wrong session. The test covers the write target but not this cross-profile eviction edge. Worth confirming the session-key is always the originating profile's key in the delayed-tap path.
2. **`_command_profile_home` is resolved before the override key is derived but only under `multiplex_profiles`.** `_resolve_profile_home_for_source(event.source)` is skipped when multiplex is off, so non-multiplexed deployments always write to `_hermes_home` — correct. Just noting the asymmetry is intentional and documented by the guard.
3. **A failed global save silently downgrades to a session override.** In `_apply_reasoning_selection`, if `_save_gateway_config_key(..., config_home=...)` returns False (e.g. the profile's config.yaml is unwritable), the code falls through to the session-override path with no signal to the user. Pre-existing behavior, but with `config_home` now pointing at an arbitrary profile the failure mode is more likely to be hit; a hint on the False branch would help diagnose a locked-down profile.

Correct fix with good regression coverage; nothing blocking.

ricardosalta:
⌾ Refreshed onto current upstream `main` (`b7d7d2929a`) without force-pushing. Current head: `6e5d6698b8`.

Relevance recheck: the flag-only `/reasoning --global` picker and delayed multiplex-profile persistence paths are still present on current `main`; upstream has not superseded this fix. The current extracted handler applies cleanly.

Validation on the refreshed head: 21 focused gateway tests passed (`test_choice_picker.py`, `test_reasoning_command.py`, `test_reasoning_config_per_model.py`), Ruff passed, and `git diff --check` is clean. I also rechecked the older review notes: the callback closes over the command-time profile-namespaced session key; the profile scope remains intentionally broader than `/fast` because reasoning `show`/`hide` choices persist per-platform display config too. Current save-failure handling explicitly reports the failed global persistence.

PR #97776, opened 2026-08-29
fix(gateway): bound and secure launchd stdio logs

tachyon-r:
## Problem

macOS launchd keeps configured stdout and stderr files open in append mode, creates them under the login-session umask, and does not rotate them. This bypasses Hermes' normal bounded logging contract and can continue writing to renamed inodes.

## Change

- set the managed launchd gateway's process umask to 077
- send launchd-owned stdout and stderr handles to `/dev/null`
- capture child stdout in `gateway.stdout.log` and timestamped stderr in `gateway.error.log`
- rotate both streams using canonical logging size and backup settings
- create and tighten live files and retained backups to 0600
- refuse symlink-following opens where `O_NOFOLLOW` is available
- apply the same capture path to the macOS detached fallback
- fall back to canonical defaults when logging rotation settings are malformed
- document where operators can find the separated raw streams

Structured gateway activity remains in `gateway.log`, so `hermes logs gateway` is unchanged.

## Verification

- focused gateway suite — 150 passed, 4 platform-specific skips
- final stderr-capture tests — 11 passed
- Ruff — passed
- Ty for `stderr_timestamp.py` — passed
- `git diff --check` — passed

## Scope and risk

The stock macOS launchd service and macOS detached fallback are affected. The 077 umask deliberately applies to the gateway process and descendants. systemd, Windows services, Nix/Home Manager launchd generation, and structured `gateway.log` behavior remain unchanged.

### DIFF
--- a/hermes_cli/gateway.py
+++ b/hermes_cli/gateway.py
@@ -3678,28 +3678,40 @@ def _timestamped_stderr_gateway_command(error_log: Path, *, external_supervisor:
         inner = [part for part in inner if part != "--replace"]
         if "--external-supervisor" not in inner:
             inner.append("--external-supervisor")
-    return [get_python_path(), "-m", "hermes_cli.stderr_timestamp", "--error-log", str(error_log), "--", *inner]
+    return [
+        get_python_path(),
+        "-m",
+        "hermes_cli.stderr_timestamp",
+        "--output-log",
+        str(error_log.with_name("gateway.stdout.log")),
+        "--error-log",
+        str(error_log),
+        "--",
+        *inner,
+    ]
 
 
 def _spawn_detached_gateway() -> bool:
     """Launch the gateway detached (launchd fallback for macOS 26+). CLI-managed nohup equivalent:
-    stdout → gateway.log, timestamped stderr → gateway.error.log, PID via gateway.pid so stop/status work.
+    raw stdout/stderr go through bounded capture, PID via gateway.pid so stop/status work.
 
     Used when launchctl can no longer bootstrap/kickstart the gateway on macOS 26+ (issue #23387). Mirrors
-    the `nohup hermes gateway run --replace` workaround but keeps it CLI-managed: stdout goes to
-    gateway.log, stderr is timestamped into gateway.error.log, and the PID is tracked via the gateway.pid
+    the `nohup hermes gateway run --replace` workaround but keeps it CLI-managed: raw stdout/stderr go through
+    the same bounded capture used by launchd, and the PID is tracked via the gateway.pid
     file that `run_gateway` writes, so stop/status/restart keep working.
     """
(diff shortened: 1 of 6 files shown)


Enough1122:

Reviewed `hermes_cli/gateway.py`, `hermes_cli/stderr_timestamp.py` + tests for PR #97776. The change correctly addresses launchd's append-only, umask-0644, non-rotating stdio handles: the plist now points `StandardOutPath`/`StandardErrorPath` at `/dev/null` and sets `Umask 63` (0o77), while the wrapper (`stderr_timestamp`) captures raw stdout/stderr into owner-only (0600), size-bounded rotating files (`gateway.stdout.log` + timestamped `gateway.error.log`). `_RotatingWriter` is solidly hardened — `O_NOFOLLOW`/`O_CLOEXEC`, `fchmod(0600)`, immediate pruning of stale backups, and `backup_count == 0` unlink path. Tests cover stdout/stderr separation, rotation bounds, backup pruning, mode tightening, and the plist assertions.

Non-blocking observations:

1. **`Umask 63` affects the whole child process, not just the log files.** The gateway process (and anything it spawns) now runs with umask 0o77, so files the gateway creates without an explicit mode become 0600 rather than 0644. For secret-bearing files that's good; but if any gateway-created file is meant to be group/shared-readable (e.g. a socket or shared config consumed by a same-group tool), it silently changes. The log files themselves are secured by explicit `fchmod`/`O_NOFOLLOW` anyway, so the umask is belt-and-braces with broader side effects — worth confirming it's intended process-wide, or narrowing it.
2. **`gateway.stdout.log` is a new file name; `gateway.log` no longer receives raw stdout on these paths.** `hermes logs gateway` and `hermes_cli/logs.py` still read the *structured* `gateway.log` (written by the logging FileHandler, unchanged), so the documented command keeps working — but operators who previously tailed `gateway.log` and saw raw stdout mixed in will now find raw prints only in `gateway.stdout.log`. A doc/message note pointing at the new file would prevent "my logs look empty" confusion.
3. **Tautological guard in `_prune_and_secure_backups`:** `candidate.name == f"{prefix}{suffix}"` where `suffix = candidate.name.removeprefix(prefix)` is always true — the real filter is `suffix.isdigit()`. Harmless, but the expression reads as if it does something; a leftover from an earlier iteration. Cosmetic.
4. **Single oversized line can exceed `max_bytes`** (rotation is between writes only) — explicitly documented in the `_RotatingWriter` docstring and consistent with `RotatingFileHandler`. Fine.
5. **`stream.tell()` on append-mode is relied on for rotation sizing.** With `buffering=0` this is reliable (position = end after each write), and the `stream.tell() and ...` short-circuit correctly skips rotation on the first write to an empty file. No issue — just noting the assumption is implicitly exercised and covered by the 32-byte rotation test.

The launchd logging contract is genuinely improved; nothing blocking.

PR #97779, opened 2026-08-29
fix(estop): honor the canonical ~/.hermes/ESTOP from profile gateways

sycamoregroupltd:
### Problem

`hermes pause` writes a sentinel at `$HERMES_HOME/ESTOP`, and `agent/estop.py` resolves that path from the **active** `HERMES_HOME` only:

```python
def _hermes_home() -> Path:
    """Resolve the active HERMES_HOME (profile-aware) at call time."""
    ...
def sentinel_path() -> Path:
    return _hermes_home() / SENTINEL_NAME
```

Profile processes launch with `HERMES_HOME=~/.hermes/profiles/<name>`, so a gateway running under a profile looks for `~/.hermes/profiles/<name>/ESTOP` and never sees the fleet-root sentinel the operator actually wrote. **The emergency stop silently fails to bind profile gateways** — the operator believes the fleet is paused while profile-dispatched work keeps running.

### Reproduction (against current `main`)

```python
root = tmp/"root"; prof = root/"profiles"/"demo"
(root/"ESTOP").write_text("paused")          # operator runs `hermes pause` at the fleet root
os.environ["HERMES_HOME"] = str(prof)        # a profile gateway's environment
from agent import estop
estop.sentinel_path()   # -> .../profiles/demo/ESTOP   (not the file the operator wrote)
estop.is_engaged()      # -> False           EXPECTED: True
```

We hit this on a fleet where `hermes pause` appeared to succeed but profile-dispatched workers kept spawning.

### Fix

Check **both** the process's own home and the canonical fleet root, for engage *and* resume, so a pause written at either location binds a profile gateway. With the change, the reproduction above returns `True`.

`tests/test_estop.py`: **25 passed** (adds 28 lines of coverage for the profile-gateway case).

### Notes

- Behaviour is unchanged when `HERMES_HOME` already *is* the canonical root — the common single-home case.
- Resume honours both locations too, so `hermes resume` cannot leave a stale sentinel that keeps a profile paused.
- Happy to adjust the approach if you'd prefer the resolution live in `hermes_constants` instead.

### DIFF
--- a/agent/estop.py
+++ b/agent/estop.py
@@ -51,24 +51,62 @@ def _hermes_home() -> Path:
         return Path(os.path.expanduser("~/.hermes"))
 
 
+def _canonical_root() -> Path:
+    """Fleet-wide Hermes root, even when this process is a profile gateway.
+
+    Profile gateways launch with HERMES_HOME=~/.hermes/profiles/<name>.
+    ``hermes pause`` from an operator seat writes ~/.hermes/ESTOP. If we
+    only inspect the profile home, the emergency stop does not bind
+    (jarvis-os/t_7b65ff88: fleet-analyst kept dispatching through pause).
+    """
+    try:
+        from hermes_constants import get_default_hermes_root
+        return Path(get_default_hermes_root())
+    except Exception:
+        return Path(os.path.expanduser("~/.hermes"))
+
+
 def sentinel_path() -> Path:
-    """Path of the ESTOP sentinel under the active HERMES_HOME."""
+    """Path of the ESTOP sentinel this process would write on `hermes pause`."""
     return _hermes_home() / SENTINEL_NAME
 
 
-def is_engaged() -> bool:
-    """Cheap check (one stat): is the global emergency stop engaged?
-
-    Fail SAFE on stat errors: if we cannot determine whether the sentinel
-    exists (permission error, transient I/O failure on HERMES_HOME), report
-    engaged. The module contract is that the pause must hold even when the
-    sentinel is unreadable — a fail-open here would silently lift an
-    operator's emergency stop exactly when the filesystem is misbehaving.
-    """
+def _candidate_sentinel_paths() -> list:
+    """Profile home first, then the fleet root if it is a different directory."""
+    primary = sentinel_path()
+    paths = [primary]
     try:
-        return sentinel_path().exists()
-    except OSError:
-        return True
+        root = _canonical_root() / SENTINEL_NAME
+    except Exception:
+        return paths
+    if not isinstance(primary, Path):
+        # Test doubles (e.g. fail-safe stat fixture) are not Path objects.
+        paths.append(root)
+        return paths
+    try:
+        if root.resolve() != primary.resolve():
+            paths.append(root)
+    except Exception:
+        if root != primary:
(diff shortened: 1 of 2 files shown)


teknium1:
Verified against `origin/main` @ `26f178e5fa`: the bug is real. `agent/estop.py:_hermes_home()` resolves via `get_hermes_home()`, which under a bound profile scope returns `~/.hermes/profiles/<name>`, so `is_engaged()` on main never sees the fleet-root `~/.hermes/ESTOP`. Using `get_default_hermes_root()` for the second candidate is the right primitive (`hermes_constants.py:183` — returns `<root>` when `HERMES_HOME` is `<root>/profiles/<name>`, and `HERMES_HOME` itself in Docker/custom layouts). Single-home installs are unaffected: `root.resolve() == primary.resolve()` collapses the candidate list to one path. The regression test covers the profile case well.

Three concerns before this merges:

1. **`disengage()` now clears the fleet-root sentinel from a profile process** (`agent/estop.py`, new `disengage` loop unlinks every candidate). That means `hermes resume` run inside any single profile lifts an operator's *fleet-wide* pause — while `engage()` remains profile-local only. The engage/resume asymmetry inverts: pausing from a profile is scoped, but resuming from a profile is global. If the operator paused the fleet at the root, one profile gateway should arguably not be able to countermand it. Suggest `disengage()` only unlink `sentinel_path()` (process-local), or at minimum call this asymmetry out as intended behavior.

2. **Test scaffolding leaked into production code**: `_candidate_sentinel_paths()` has `if not isinstance(primary, Path): # Test doubles ... are not Path objects`, and `disengage()`/`get_state()` catch `AttributeError` for the same reason. Production paths shouldn't special-case test fixtures — adapt the fail-safe-stat fixture (make its double subclass `Path` or patch at a different seam) rather than widening exception handling in shipping code.

3. **Hot-path cost**: the module contract ("single `os.stat` so callers may run it every tick") goes from 1 stat to up to 2 `exists()` + 2 `resolve()` calls (each `resolve()` also hits the filesystem) on every scheduler tick and gateway message. `get_default_hermes_root()` is memoized, but the candidate list and `resolve()` comparison are rebuilt per call. Cheap fix: compute the deduped candidate pair once per (native_home, env_home) the same way `hermes_constants` memoizes the root.

Verdict: right diagnosis, right root primitive, needs-work on the resume semantics (#1) and the test-double leakage (#2) before merge.

kshitijk4poor:
Merged via #99653 — your commits were cherry-picked with authorship preserved (rebase-merge), plus one follow-up commit dropping a redundant test-double branch and updating the module docstring.

Your fix closes a real operational hazard: a gateway launched under `HERMES_HOME=~/.hermes/profiles/<name>` never saw the fleet-root sentinel `hermes pause` actually wrote, so the emergency stop silently failed to bind profile gateways. The dual-candidate sentinel check (process home + canonical fleet root, honoring `hermes resume` from either seat) shipped exactly as you designed it.

Validated during salvage: bug reproduced on origin/main pre-fix, 25/25 estop tests, 17/17 E2E probe assertions (profile binding, Docker layout, corrupt-sentinel fail-safe, reason preference), and the design asymmetry documented in the PR body.

Thanks for the contribution!

PR #84010, opened 2026-08-11
fix(delegation): treat literal 'auto' model/provider as inherit-from-parent

pittosporum-seu:
## What does this PR do?

Normalizes the literal `"auto"` sentinel in `delegation.model` / `delegation.provider` to the documented *inherit from parent* behavior, instead of treating it as an explicit override that reaches the wire.

`cli-config.yaml.example` documents `empty = inherit parent` for both keys, but users and setup flows commonly write `"auto"` with the same intent. Without normalization:

- `model="auto"` is passed verbatim to the child `AIAgent` and rejected by the API (HTTP 401 `Model auto is not supported`) — **every subagent spawn fails** (non-retryable).
- `provider="auto"` reaches `resolve_runtime_provider(requested="auto")`, matches no configured provider, and silently reroutes subagents through the fallback chain — spending an unrelated provider's balance.

The fix adds `_normalize_auto_sentinel()` (maps literal `"auto"`, case-insensitive/stripped, to `None`) and applies it to both axes at the top of `_resolve_delegation_credentials()`. Explicit real overrides pass through untouched. This mirrors the auxiliary-path sentinel contract in `agent/auxiliary_client.py` and the cron-job normalization in `cron/scheduler.py` (#83596 / companion PR).

## Related Issue

Fixes #84007

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `tools/delegate_tool.py`:
  - new `_normalize_auto_sentinel()` helper
  - `_resolve_delegation_credentials()`: normalize `delegation.model` / `delegation.provider` before override branching, so `"auto"` takes the documented inherit-from-parent path
- `cli-config.yaml.example`: document `empty or "auto" = inherit parent` for both keys
- `tests/tools/test_delegate_auto_sentinel.py`: 9 new tests (helper unit tests, inherit behavior, case-insensitivity, sentinel == empty equivalence, explicit-override passthrough, mixed pin/auto)

## How to Test

1. Configure `delegation: {model: auto, provider: auto}` and delegate any task — before the fix every child fails with `HTTP 401: Model auto is not supported`; after the fix children inherit the parent's model/provider.
2. `pytest tests/tools/test_delegate_auto_sentinel.py -q` — 9 passed.
[...]

PR #96051, opened 2026-08-27
feat(plugins): let a pre_llm_call hook route a turn to another model

juniperbevensee:
## What does this PR do?

It lets a `pre_llm_call` plugin **return a routing decision that the runtime actually carries out**.

Today that hook can change the *text* of a turn (context injection) but not *which model answers it*. By the time it fires, `restore_primary_runtime` has already fixed the turn's model and rebuilt the client. So a plugin can decide "this turn should run somewhere cheaper/faster" and then has no way to act on the decision.

A hook result may now carry a `route` key alongside the existing `context` key:

```python
def route_turn(user_message, model, **kwargs):
    if is_mechanical(user_message):
        return {"route": {"model": "deepseek/deepseek-v3.2",
                          "provider": "openrouter"}}
    return None  # no route -> the configured model answers

def register(ctx):
    ctx.register_hook("pre_llm_call", route_turn)
```

`model` is required; `provider`, `base_url`, `api_key`, and `api_mode` are optional and forwarded to the agent's existing `switch_model`.

**No router, classifier, tiers, or heuristics land in core.** There are no new config keys and no default-on behavior. Core gains one thing: the ability to *carry out* a decision a plugin already made. A complete opt-in example plugin is included in the docs — it lives in `~/.hermes/plugins/`, not in this tree.

### Why this shape, and not a router

`#43534` proposed a router in core and was closed by @teknium1 with:

> If you do want it, please create a plugin. Happy to support adding to the plugin interface to make this work as one.

This PR is that plugin-interface addition and nothing else. The decision stays in plugin territory, which is where `#43534` concluded it belongs.

I want to be upfront about the tension: `#61371` was closed `not_planned` under a standing `delegation-model-routing` policy — *"Hermes does not accept per-task model/provider routing, model tiers, or complexity-based automatic model selection"* — and `424e9f36b0` deliberately removed the old `smart_model_routing` feature. Nothing here re-adds any of that to core. If the intent is that the policy covers the *plugin interface* too, not just core behavior, then this should be closed and I'd rather hear that than have you carry the review.

### Extend, don't duplicate — what I checked first

- `ctx.llm` (`agent/plugin_llm.py`) is the host-owned LLM facade for a plugin's **own out-of-band** calls. It cannot change the model of the agent's turn.
[...]

PR #92905, opened 2026-08-23
feat(model): enforce profile-wide hard context ceiling

Forser:
## Summary

Add `model.max_context_length` as a profile-wide hard ceiling on Hermes's effective operating context.

```text
effective_context = min(invocation_pre_cap_context, profile_max_context_length)
```

The ceiling applies when `model.max_context_length` is valid and configured. It can only lower an invocation's resolved context; it never inflates a smaller model or provider limit.

This extends the `max_context_length` ceiling concept proposed in #70242 from compression-threshold scope to coherent runtime policy, terminal dispatch enforcement, and output reservation.

## Behavior

- `model.max_context_length` is profile-wide and survives model/provider changes.
- `model.context_length` remains an exact per-model pre-cap override, but its effective value is still subject to the profile ceiling.
- `model.max_tokens` remains the per-response output cap.
- Valid ceilings are integers at or above Hermes's existing 64K operating floor. Invalid values are ignored with a ceiling-specific diagnostic.
- Provider/model-resolved pre-cap values remain independently preserved for capability metadata, cache persistence, switching, fallback, and restore-primary behavior.
- Persistent context capability caches remain raw and are not contaminated by profile policy.

## Implementation highlights

- Centralizes strict ceiling validation, effective-limit calculation, output reservation, canonical final-request budgeting, and typed local refusal.
- Keeps active pre-cap and effective context coherent across initialization, explicit context overrides, `/model` switching, provider changes, fallback activation, rollback, and restore-primary.
- Applies each invocation's own pre-cap to auxiliary sync/async calls and MoA reference calls while sharing the profile-wide ceiling.
- Recomputes compression thresholds and budgets from the same effective context and resolved output reservation used at dispatch.
- Enforces the limit on final provider-bound payloads, including Relay-transformed payloads, before physical OpenAI-compatible, Anthropic, Bedrock, Codex, and streaming dispatch.
- Covers iteration-limit summary calls and treats `ContextCeilingExceeded` as a terminal local refusal: no provider call, retry, credential refresh, fallback, or API-call accounting.
- Keeps gateway cached-agent reuse, profile isolation, context-reference admission, TUI/Desktop paths, and status/model reporting aligned with the effective context.
- Preserves plugin compatibility without model-specific ceiling special cases.

## Validation

Fresh bounded processes after the final upstream integration:

- 21/21 Round 5 ownership, Relay/streaming, rollback, restore-primary, and compatibility tests passed.
[...]

PR #97780, opened 2026-08-29
feat(runtime): immutable ResolvedRuntime + central client factory; fixes extra_headers loss in fallback paths

### LINKED ISSUE #88463: [Bug]: Custom provider extra_headers dropped during auxiliary auto-resolution and CLI model probing
### Description

When using custom providers (e.g. Bifrost, LiteLLM, or other OpenAI-compatible gateways) that require custom authentication or routing headers (such as `x-bf-vk` or tenant tokens via `extra_headers`), several downstream subsystem paths drop or fail to thread these headers, resulting in `HTTP 401: Unauthorized` / missing required header errors while main chat requests succeed.

Specifically, two distinct paths fail to forward `extra_headers`:

---

### Bug 1: Auxiliary `auto` resolution drops `extra_headers`
* **Location:** `agent/auxiliary_client.py` (`_resolve_auto_route`)
* **Behavior:** When auxiliary tasks (such as `title_generation`, `compression`, etc.) are configured with `provider: auto`, Hermes attempts to use the main provider's endpoint. When the main provider is a named custom provider (`custom:<name>`), `_resolve_auto_route` resolves the provider client via `resolve_provider_client`, but fails to carry forward the custom provider's configured `extra_headers`.
* **Consequence:** Tasks like automatic session title generation fail with `HTTP 401: virtual key is required. Provide a virtual key via the x-bf-vk header` (or equivalent auth errors) even though the primary agent loop communicates with the endpoint without issue.

---

### Bug 2: Interactive CLI model discovery (`_model_flow_named_custom`) does not forward `extra_headers`
* **Location:** `hermes_cli/model_setup_flows.py` (`_model_flow_named_custom`)
* **Behavior:** When running `hermes model` and selecting a named custom provider, `_model_flow_named_custom` attempts to probe `/v1/models` via `fetch_api_models(api_key, base_url, **fetch_kwargs)`. However, `_model_flow_named_custom` only extracts `api_mode` and does not extract or pass `provider_info.get("extra_headers")` into `fetch_api_models`.
* **Contrast:** In `hermes_cli/model_switch.py` (the TUI/runtime switcher), `extra_headers` are properly extracted and passed via `cached_fetch_api_models(..., headers=grp.get("extra_headers"))`. The interactive CLI setup wizard omitted this parameter.
* **Consequence:** Interactive model selection fails to probe `/v1/models` against header-gated endpoints, returning 401s during live catalog discovery.

---

### Suggested Fixes
1. **In `agent/auxiliary_client.py` (`_resolve_auto_route`):** Ensure `extra_headers` defined on the active `custom_providers` entry are merged and forwarded when constructing the client for `auto` auxiliary tasks.
2. **In `hermes_cli/model_setup_flows.py` (`_model_flow_named_custom`):** Extract `extra_headers = provider_info.get("extra_headers")` (normalized via `normalize_extra_headers`) and pass `headers=extra_headers` to `fetch_api_models()`.

pittosporum-seu:
## What does this PR do?

Fixes provider drift during startup, delegation, fallback, model switching, and
transport recovery. A resolved route previously became split across mutable
agent fields and independently rebuilt SDK clients, losing headers, timeout,
API mode, TLS, query, or provider-specific metadata on later requests.

The architecture has one complete value and one installation boundary:

1. `ResolvedRuntime` is an immutable, mapping-compatible provider contract.
2. `build_client_bundle(runtime)` constructs the wire client without mutating a live agent.
3. `AIAgent.install_runtime(bundle)` atomically installs identity and client state under the existing lock, then retires the old client.
4. Startup, auxiliary resolution, subagent inheritance, fallback, credential rotation, primary restoration, and model switching carry this complete contract through their existing policy flows.

The follow-up fixes the remaining boundaries exposed by the new immutable
representation. Each child owns a fresh client; same-route settings survive
handoffs, while an endpoint change resolves destination settings without leaking
the previous route's headers, query, or TLS configuration. Bedrock and MoA retain
their dedicated builders; this PR does not claim those special wires have been
fully migrated.

## Related Issue

Fixes #88463.

Related: #84010 (subagent `auto` inheritance). This updates the existing #97780
architecture PR rather than creating a competing fix.

## Type of Change

- [x] Refactor (no behavior change outside the repaired paths)
- [x] Tests (adding or improving test coverage)

## Changes Made

- `agent/runtime_bundle.py`: immutable `ResolvedRuntime`, `ClientBundle`, recursive mutable configuration copies, and central transport-aware client construction.
- `agent/runtime_routes.py`: complete runtime handoffs, construction metadata, destination-specific route configuration, and separately owned child clients.
- `hermes_cli/runtime_provider.py` and `runtime_provider_custom.py`: freeze resolver results and retain named-provider TLS and request metadata.
- `tools/delegate_tool_config.py` and `delegate_tool.py`: inherit the live parent runtime, normalize `auto`, accept frozen mappings, and preserve explicit-over-provider request override precedence.
- `agent/agent_init.py` and `run_agent.py`: accept a resolved runtime directly and preserve it in primary snapshots.
- `agent/client_lifecycle.py`: runtime-aware Anthropic request reuse and transactional credential rotation/refresh; failed construction retains the previous client/runtime.
- `agent/auxiliary_client.py`, `anthropic_adapter.py`, and `chat_completion_helpers.py`: retain headers, CA, query, and timeout through fallback and later request-local reconstruction.
- `agent/agent_runtime_helpers.py`: central construction during recovery, restoration, and model switching.
- `tui_gateway/server.py` and `hermes_cli/cli_agent_setup_mixin.py`: accept immutable mapping-compatible resolver results without mutation.
- Regression tests cover immutable overrides, nested delegation, fallback-to-child transitions, request reuse, rotation rollback, transport propagation, and client ownership. Existing fixtures now exercise the real installation contract; Windows test paths and encoding are explicit.
- `website/docs/developer-guide/provider-runtime.md`: document runtime handoffs and ownership.

### Reproduction and Integration Coverage

The follow-up defect tests initially produced nine failures on the architecture
head before these fixes. They exercise the actual delegation resolver and
request-client lifecycle: nested frozen overrides previously raised a
mappingproxy copy error or disappeared; required Anthropic headers were dropped;
rotation left stale runtime state; and literal `auto` escaped inheritance.

Real OpenAI and Anthropic SDK integration tests use temporary `HERMES_HOME`
configuration and offline httpx transports. They construct parents, children,
and grandchildren; activate fallback through the real resolver; rotate
credentials; inspect outgoing request headers/query; and verify separate client
ownership and route isolation. No live model calls or credentials are needed.

### Focused Regression Command

Use the project development venv, or set `HERMES_PYTHON` to its Python executable
when the checkout does not contain a venv:

```bash
bash scripts/run_tests.sh tests/tools/test_delegate*.py \
  tests/agent/test_runtime_bundle.py tests/agent/test_runtime_install.py \
  tests/agent/test_runtime_boundary_preservation.py \
  tests/agent/test_anthropic_adapter.py tests/agent/test_anthropic_request_client_reuse.py \
  tests/agent/test_request_client_reuse.py tests/agent/test_auxiliary_client.py \
  tests/agent/test_auxiliary_client_anthropic_custom.py tests/agent/test_auxiliary_client_ssl_verify.py \
  tests/agent/test_auxiliary_client_azure_foundry.py tests/agent/test_auxiliary_user_default_headers.py \
  tests/agent/test_moa_switch_api_mode.py tests/run_agent/test_primary_runtime_restore.py \
  tests/run_agent/test_openai_client_lifecycle.py tests/run_agent/test_credential_rotation_route_settings.py \
  tests/run_agent/test_fallback_credential_isolation.py tests/run_agent/test_switch_model_rollback.py \
  tests/run_agent/test_env_credential_turn_refresh.py tests/run_agent/test_create_openai_client_reuse.py \
  tests/run_agent/test_request_client_reuse_abort_races.py tests/run_agent/test_credential_pool_interrupt.py \
  tests/hermes_cli/test_runtime_provider_resolution.py tests/hermes_cli/test_runtime_transport_precedence.py \
  tests/test_profile_isolation_runtime.py -j 4 -q
```

### Validation Record

- Focused regression on Windows: **43 files, 738 passed, 0 failed, 1 skipped**. The skipped test asserts POSIX permission bits, unsupported on Windows.
- Ruff `0.15.10` on changed Python files: passed.
- `scripts/check-windows-footguns.py` on the staged Python files: passed.
- `scripts/check_compat_pointers.py`: passed.
- `git diff --check`: passed.
- Earlier architecture-head validation recorded 179 focused tests, 637 TUI gateway tests, and a 41-test switch-boundary refinement rerun. These are historical receipts, not additional tests counted in the latest 738-pass run.

The entire repository suite and a live provider/gateway smoke test have not been
run for this follow-up. The active Hermes installation was not deployed or
restarted. The branch's original base is `245e48008fa814b3251f50755eb656bd9fb86cb1`;
this is not a claim that the branch is synchronized with today's upstream main.

### Code

- [x] I've read the Contributing Guide.
- [x] Commit messages follow Conventional Commits.
- [x] Existing related work was checked; this remains in the original #97780 PR.
- [x] The changes are scoped to provider runtime propagation and its regressions.
- [x] Focused related tests pass; the full repository suite is not claimed.
- [x] Bug regressions and offline real SDK integration tests are included.
- [x] Tested on Windows.

- [x] Provider runtime developer documentation and relevant docstrings are updated.
- [x] `cli-config.yaml.example`: N/A, no new config keys.
- [x] `CONTRIBUTING.md` / `AGENTS.md`: existing boundary and invariant guidance retained; detailed architecture changes are in the developer guide.
- [x] Cross-platform impact considered; Windows path, encoding, and static checks covered. Linux/macOS execution remains for CI.
- [x] Tool descriptions/schemas: N/A, no new tool or schema fields.

### DIFF
--- a/agent/agent_init.py
+++ b/agent/agent_init.py
@@ -682,7 +682,6 @@ def _print_key_banner(key, label: str, warn_missing: bool = False) -> None:
 
 def _init_anthropic_client(agent, api_key, base_url, _provider_timeout):
     """anthropic_messages: native Anthropic SDK (or AnthropicBedrock for Bedrock+Claude)."""
-    from agent.anthropic_adapter import build_anthropic_client
     from agent.anthropic_credentials import resolve_anthropic_token
     agent.client = None
     agent._client_kwargs = {}
@@ -718,17 +717,37 @@ def _init_anthropic_client(agent, api_key, base_url, _provider_timeout):
                 _mm_exc,
             )
 
-    agent.api_key = effective_key
-    agent._anthropic_api_key = effective_key
     # OAuth only for native Anthropic: third-party anthropic_messages providers must never
     # trip OAuth paths — those inject Claude-Code identity headers → 401/403.
     # Only mark the session as OAuth-authenticated when the token genuinely belongs to native Anthropic.
     # Third-party providers (MiniMax, Kimi, GLM, LiteLLM proxies) that accept the Anthropic protocol must
     # never trip OAuth code paths — doing so injects Claude-Code identity headers and system prompts that
     # cause 401/403 on their endpoints. See #1739.
-    from agent.anthropic_credentials import _is_oauth_token as _is_oat
-    agent._is_anthropic_oauth = _is_oat(effective_key) if (_is_native_anthropic and isinstance(effective_key, str)) else False
-    agent._anthropic_client = build_anthropic_client(effective_key, base_url, timeout=_provider_timeout)
+    # Build the complete wire client before touching the live agent.  The
+    # provider-entry header layer must travel with the runtime; rebuilding a
+    # fallback from only key + URL is what originally dropped gateway auth.
+    from agent.runtime_bundle import ResolvedRuntime, build_client_bundle
+    runtime_fields = {
+        "provider": agent.provider,
+        "requested_provider": agent.requested_provider,
+        "model": agent.model,
+        "api_mode": agent.api_mode,
+        "api_key": effective_key,
+        "base_url": base_url or "",
+    }
+    with suppress(Exception):
+        from hermes_cli.config import (
+            get_compatible_custom_providers, get_custom_provider_extra_headers,
+            load_config_readonly,
+        )
+        _headers = get_custom_provider_extra_headers(
+            base_url or "", get_compatible_custom_providers(load_config_readonly())
+        )
+        if _headers:
+            runtime_fields["extra_headers"] = _headers
+    runtime = ResolvedRuntime.from_mapping(runtime_fields)
+    bundle = build_client_bundle(runtime, timeout=_provider_timeout)
+    agent.install_runtime(bundle, reason="agent_init")
     if not agent.quiet_mode:
         print(f"🤖 AI Agent initialized with model: {agent.model} (Anthropic native)")
         _print_key_banner(effective_key, "token")
@@ -914,14 +933,36 @@ def _init_openai_client(agent, api_key, base_url, fallback_model, _provider_time
         if agent.provider == "bedrock" and "bedrock-mantle." in str(client_kwargs.get("base_url", "")):
             raise
 
-    agent._client_kwargs = client_kwargs  # stored for rebuilding after interrupt
+    agent._client_kwargs = client_kwargs  # temporary policy input; install_runtime owns the committed copy
     _apply_openai_header_policy(agent, client_kwargs)
     agent.api_key = client_kwargs.get("api_key", "")
     agent.base_url = client_kwargs.get("base_url", agent.base_url)
     try:
         from agent.ssl_guard import verify_ca_bundle
         verify_ca_bundle()
-        agent.client = agent._create_openai_client(client_kwargs, reason="agent_init", shared=True)
+        if agent.provider == "bedrock":
+            # Bedrock Mantle keeps its SigV4-aware path until that special wire
+            # is migrated behind runtime_bundle.
+            agent.api_key = client_kwargs.get("api_key", "")
+            agent.base_url = client_kwargs.get("base_url", agent.base_url)
+            agent.client = agent._create_openai_client(client_kwargs, reason="agent_init", shared=True)
+        else:
+            from agent.runtime_bundle import ResolvedRuntime, build_client_bundle
+
+            runtime = ResolvedRuntime.from_mapping({
+                "provider": agent.provider,
+                "requested_provider": agent.requested_provider,
+                "model": agent.model,
+                "api_mode": agent.api_mode,
+                **client_kwargs,
+            })
+            bundle = build_client_bundle(
+                runtime,
+                openai_builder=lambda kwargs: agent._create_openai_client(
+                    kwargs, reason="agent_init", shared=True, runtime=runtime
+                ),
+            )
+            agent.install_runtime(bundle, reason="agent_init")
         if not agent.quiet_mode:
             print(f"🤖 AI Agent initialized with model: {agent.model}")
             if base_url:
@@ -2079,7 +2120,7 @@ def _snapshot_primary_runtime(agent):
     # Per-turn restoration snapshot: after a fallback, the next turn restores these so the
     # preferred model gets a fresh attempt.
     _cc = agent.context_compressor
-    agent._primary_runtime = {
+    runtime = {
         "model": agent.model,
         "provider": agent.provider,
         "requested_provider": agent.requested_provider,
@@ -2098,13 +2139,20 @@ def _snapshot_primary_runtime(agent):
         "compressor_provider": getattr(_cc, "provider", agent.provider),
         "compressor_context_length": _cc.context_length,
         "compressor_threshold_tokens": _cc.threshold_tokens,
+        "compressor_api_mode": getattr(_cc, "api_mode", agent.api_mode),
+        "runtime_capabilities": dict(getattr(agent, "runtime_capabilities", {}) or {}),
+        "reasoning_config": dict(getattr(agent, "reasoning_config", {}) or {}) or None,
     }
     if agent.api_mode == "anthropic_messages":
-        agent._primary_runtime.update({
+        runtime.update({
             "anthropic_api_key": agent._anthropic_api_key,
             "anthropic_base_url": agent._anthropic_base_url,
             "is_anthropic_oauth": agent._is_anthropic_oauth,
         })
+    # Keep this historical turn snapshot mutable for plugin/test compatibility.
+    # The authoritative live route is the immutable ``agent._resolved_runtime``
+    # installed by the client bundle.
+    agent._primary_runtime = runtime
 
 
 def _init_usage_state(agent):
--- a/agent/agent_init.py
+++ b/agent/agent_init.py
@@ -738,13 +738,14 @@ def _init_anthropic_client(agent, api_key, base_url, _provider_timeout):
     with suppress(Exception):
         from hermes_cli.config import (
             get_compatible_custom_providers, get_custom_provider_extra_headers,
-            load_config_readonly,
+            load_config_readonly, apply_custom_provider_tls_to_client_kwargs,
         )
         _headers = get_custom_provider_extra_headers(
             base_url or "", get_compatible_custom_providers(load_config_readonly())
         )
         if _headers:
             runtime_fields["extra_headers"] = _headers
+        apply_custom_provider_tls_to_client_kwargs(runtime_fields, base_url or "")
     runtime = ResolvedRuntime.from_mapping(runtime_fields)
     bundle = build_client_bundle(runtime, timeout=_provider_timeout)
     agent.install_runtime(bundle, reason="agent_init")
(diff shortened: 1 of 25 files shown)


andrexibiza (commented):
Reviewed exact head `42fe0f3a51242e7bc4b9ba51c4d108961abd5f58` against the PR base `aff5125f8edf5095aef5d3d79bbbb101c95b9413`, then reconciled it with current `main@a2e19d484cb5591df8dafe667c93345b62d9bf06`. I read the full behavior-critical runtime surfaces (`runtime_bundle`, auxiliary resolution, startup/install, switch/restore, fallback activation, Anthropic construction, TUI resume), the new contract tests, #88463/#88474, the open switch/fallback interlocks, and the exact-head workflow state. There were no prior reviews or unresolved review threads on this head.

There is a lot worth keeping here. `ResolvedRuntime` is actually immutable down into nested mappings, the bundle factory gives `extra_headers` the right precedence over defaults, the named-custom auxiliary path now carries header/TLS state through one typed value, and the TUI follow-up correctly repaired persisted override application with `with_updates()`. Exact-head CI/Docker/Nix are green (`CI` run 33244907628, Docker 33244907378, Nix 33244907376), and the 27-job CI run has no failed/cancelled job. This is a meaningful improvement over the current duplicated resolver shape.

I cannot clear the current head yet. I found two concrete merge blockers, one of which is the exact other side of the header-loss shape this PR is intended to eliminate.

### P1 — `anthropic_messages` fallback still reconstructs the client outside the bundle and drops the required entry headers

The new named-custom resolver is correct in isolation: `agent/auxiliary_client.py` builds the `ResolvedRuntime`, `build_client_bundle()` passes `extra_headers` as `default_headers` to `build_anthropic_client()`, and `resolve_provider_client()` wraps `bundle.anthropic_client` in `AnthropicAuxiliaryClient`. The new `test_anthropic_wire_bundle_merges_runtime_headers_through_builder` pins exactly that seam.

But the actual agent fallback activation does not install that resolved client/bundle. At this exact head, `agent/chat_completion_helpers.py::try_activate_fallback()` reaches `fb_api_mode == "anthropic_messages"`, takes only `fb_client.api_key` / `fb_base_url`, and then does a second construction:

```python
agent._anthropic_client = build_anthropic_client(
    effective_key, agent._anthropic_base_url, timeout=_fb_timeout,
)
```

No `default_headers` are supplied. `build_anthropic_client()` explicitly supports and merges `default_headers`, so a named custom Anthropic-compatible fallback whose `extra_headers` are required for WAF/auth/routing gets the header-correct client from `resolve_provider_client()`, discards it, and recreates a header-less native client at activation. That is the same defect class as the bug this PR is closing, on the fallback-to-Anthropic projection rather than the auxiliary-resolution projection.

This also falsifies the PR's current structural claim that “no construction path exists outside the factory.” There are still live direct constructors/mutation owners in `try_activate_fallback()`, `switch_model()`, and `restore_primary_runtime()`; in particular, the exact-head `switch_model()` still has its manual old-state snapshot/rollback path and direct Anthropic client construction even though the phase-3 commit message says the 14-field switch rollback was replaced by runtime installation.

**Required fix:** make fallback activation consume/install the already-resolved runtime/bundle rather than reconstructing the Anthropic client from ambient fields, or otherwise prove that the complete header/TLS/auth contract is preserved at that second constructor. Add a real regression that configures a named custom `anthropic_messages` fallback with a required `extra_headers` value, activates it through `try_activate_fallback()`, and verifies the installed request client carries that header. Then exercise startup → switch → fallback → restore against the same runtime contract so the remaining construction owners cannot silently diverge. If this PR intentionally stops short of migrating switch/restore, narrow the “single factory / all paths” claims accordingly rather than using a stronger invariant than the code establishes.

This interlocks directly with #94036 (native-compaction capability across switch/fallback/restore), #92905 (transactional context ceiling across the same state transitions), and #96051 (turn-scoped routing through switch/restore). Those are adjacent, not duplicates. If this PR centralizes those mutation boundaries now, their capability/policy state has to become part of the installed runtime contract rather than being overwritten by a mechanically narrower swap.

### P1 — `Fixes #88463` would close an issue whose second reported path this PR does not repair

#88463 is explicitly a two-path bug:

1. named-custom auxiliary auto-resolution drops `extra_headers`; and
2. `hermes_cli/model_setup_flows.py::_model_flow_named_custom()` probes `/models` without those headers, so header-gated providers still 401 during interactive model discovery.

This PR fixes/generalizes (1), but its 14-file diff does not touch `hermes_cli/model_setup_flows.py` or add a model-probe regression. #88474 by `liuhao1024` does contain that second fix: it passes `provider_info["extra_headers"]` into `fetch_api_models(..., headers=...)` and has focused coverage for the present/absent-header cases.

As written, merging #97780 with `Fixes #88463` can close the canonical issue while the CLI probe defect remains live on `main`. The contributor/supersession relationship is also more precise than “this replaces #88474”: #97780 structurally supersedes/generalizes #88474's **auxiliary construction half**, while #88474 remains **complementary and unsuperseded** for the interactive model-probe half.

**Required fix:** either carry the model-probe fix + regression into this PR while preserving explicit credit to `liuhao1024` / #88474, or change the issue relationship so #97780 does not close #88463 and leave #88474 (or a clean successor) owning path (2). Please do not collapse the two contributions into a duplicate/superseded label when only one half is actually superseded.

### Current-main / release proof

Current `main@a2e19d484cb5591df8dafe667c93345b62d9bf06` is one commit beyond this PR's base; that commit changes `agent/prompt_builder.py` and `tests/agent/test_prompt_builder.py`, so I found no direct FILE-LIST collision with #97780. The current head is mergeable and its exact-head CI/Docker/Nix receipts are green. After the blockers above are repaired, the final candidate still needs to be refreshed/recomposed on current main and re-run through those exact-object gates; the present green receipts prove `42fe0f3a...` on its current base, not the eventual merge object.

The central-runtime direction is the right kind of fix for this class: carry the whole route contract once, then make later boundaries consume that typed object rather than reconstruct configuration from ambient fields. The current patch gets that right at resolution and startup; closing the two remaining gaps above would make the claim match the mechanism. Strong work getting this much of a very drift-prone surface under one value contract.

pittosporum-seu:
Both blockers addressed — new head `56e8af9ea4`:

**P1-1 (anthropic fallback reconstruction)** — `try_activate_fallback()` now installs the resolved bundle's Anthropic client instead of rebuilding from ambient fields (`99c38ddf57`). The fallback path can no longer drop the entry `extra_headers` on the `anthropic_messages` projection. Added the requested regression: a named-custom `anthropic_messages` fallback with a required header, activated through the real `try_activate_fallback()` path, asserting the installed client carries it.

**P1-1 (declaration vs mechanism)** — chose to **narrow the claim**: the PR description now states resolution + startup + fallback activation route through the factory; `switch_model()`/`restore_primary_runtime()` migration is explicitly a follow-up (with a note that #94036/#92905/#96051 interlocks must be part of the installed-runtime contract when that happens).

**P1-2 (#88463 two-path)** — carried the interactive model-probe fix from #88474 into this PR (`hermes_cli/model_setup_flows.py`: probes now pass `extra_headers` via `fetch_api_models(..., headers=...)`, omitted when absent), with present/absent-header regression cases. Credit to @liuhao1024 preserved in the PR body and as co-authorship in `99c38ddf57`. `Fixes #88463` now covers both reported paths.

Also rebased onto current `main` (merge `23c480f6a6`), which pulled in the desktop stop/interrupt fixes — two follow-up commits (`1b47277067`, `5c3bfb6da8`) repair a stop-vs-marker race those exposed in our TUI paths.

Exact-head CI on the new head: 27/27 green including Python tests/e2e and Windows-only lanes.

pittosporum-seu:
Friendly ping — both blockers from the last review round were addressed in `56e8af9` (Sep 1: anthropic fallback reconstruction installs the resolved bundle client + regression test through the real `try_activate_fallback()` path). CI is green since. Please let me know if there's anything else you'd like changed before merge.

PR #97783, opened 2026-08-29
fix(cron): accept bare duration units like 'hour' in schedule parsing

salch-cred:
Fixes the schedule-parser/UI mismatch where `"every hour"` (advertised as a valid example) errors with `Invalid duration: 'hour'`, while `"every 1h"` works.

## Root cause

`cron/jobs.py` `parse_duration()` matched `re.match(r'^(\d+)\s*(...)$')` — a leading digit was mandatory. `parse_schedule("every hour")` strips `"every "` → `"hour"` → `parse_duration("hour")` failed the regex and raised. Bare-unit phrases like `"hour"` / `"day"` had no way to parse.

## Fix

Made the leading number optional, defaulting to `1` when absent:

```python
match = re.match(r'^(\d*)\s*(m|min|...|hour|hours|d|day|days)$', s)
value = int(match.group(1)) if match.group(1) else 1
```

Now `"hour"` → 60, `"day"` → 1440, `"min"` → 1, and existing forms (`"30m"`, `"2h"`, `"1d"`, `"every 2h"`) are unchanged.

## Validation

- `parse_duration("hour")` → 60; `"day"` → 1440; `"min"` → 1.
- `parse_schedule("every hour")` → `{kind: interval, minutes: 60}`.
- Existing `"30m"`, `"2h"`, `"1d"`, `"every 2h"`, `"every 30m"` all unchanged.

### DIFF
--- a/contributors/emails/salch-cred@users.noreply.github.com
+++ b/contributors/emails/salch-cred@users.noreply.github.com
@@ -1,0 +1,1 @@
+salch-cred
--- a/cron/jobs.py
+++ b/cron/jobs.py
@@ -716,13 +716,14 @@ def parse_duration(s: str) -> int:
         "30m" → 30
         "2h" → 120
         "1d" → 1440
+        "hour" → 60 (bare unit, no leading number)
     """
     s = s.strip().lower()
-    match = re.match(r'^(\d+)\s*(m|min|mins|minute|minutes|h|hr|hrs|hour|hours|d|day|days)$', s)
+    match = re.match(r'^(\d*)\s*(m|min|mins|minute|minutes|h|hr|hrs|hour|hours|d|day|days)$', s)
     if not match:
         raise ValueError(f"Invalid duration: '{s}'. Use format like '30m', '2h', or '1d'")
     
-    value = int(match.group(1))
+    value = int(match.group(1)) if match.group(1) else 1
     unit = match.group(2)[0]  # First char: m, h, or d
     
     multipliers = {'m': 1, 'h': 60, 'd': 1440}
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed `cron/jobs.py` + a contributor email marker for PR #97783. The change is minimal and correct: `^(\d+)\s*` → `^(\d*)\s*` with `int(match.group(1)) if match.group(1) else 1` makes bare units (`hour`, `day`, and by extension `h`/`m`/`d`) resolve to a quantity of 1. This fixes the advertised `"every hour"` schedule that previously raised `Invalid duration: 'hour'`.

Non-blocking observations:

1. **Bare shorthand units are now valid too, which is broader than the title implies.** `"h"`, `"m"`, `"d"` (and single letters in general) now parse as 1 hour/minute/day. Consistent with the "bare unit means 1" rule, so this is fine — just noting the PR title says "hour"/"day" while the regex opens up every unit letter.
2. **The ValueError message still only lists `'30m', '2h', or '1d'`.** A user typing a bare unit that *doesn't* match (e.g. `"sec"`, `"minute "` with trailing content, `"every 2h "` mishandled elsewhere) gets guidance that doesn't mention bare units. Worth appending `or a bare unit like 'hour'` to the error text.
3. **Minor:** the `(m|min|mins|minute|minutes|...)` alternation is still greedy-first on `m`, so `match.group(2)[0]` (first char) is a reliable unit discriminator — no change needed, just confirming the existing design still holds with the now-optional digit.

Correct fix, cleanly scoped; nothing blocking.

salch-cred:
Good catch on the error text — since bare units are now valid, the message should say so. Updated it to mention 'a bare unit like hour (defaults to 1)'. The broadened shorthand (h/m/d now parse as 1) is intentional and consistent with the 'bare unit means 1' rule, so I left that as-is.

teknium1:
Merged via PR #98252 — your commits were cherry-picked onto current main with your authorship preserved in git log, plus a small test-coverage follow-up on top. Thanks for the fix!

salch-cred:
Thanks @teknium1 — really appreciate you salvaging this, preserving the authorship, and adding the test coverage on top. Was a good catch from the Discord thread on 'every hour' failing. Glad it's in.

PR #97789, opened 2026-08-29
fix(desktop): mark previous reply spoken before submitting the next voice turn

### LINKED ISSUE #97788: Desktop voice conversation re-speaks the previous reply after the next voice turn is submitted
## Summary

In the Desktop voice conversation loop (hands-free mode, the **End** button session), the **previous** assistant reply is spoken again right after the user's next utterance is submitted, while the model is still thinking. Playback of the stale reply is cut off as soon as the new reply starts streaming, so each turn begins with a few seconds of the old answer.

## Environment

- Hermes Desktop (macOS 15, Apple Silicon), repo at `ee742fe1bc` (2026-08-29); `origin/main` has the same code in the affected file
- `tts.provider: edge`, `stt.provider: local` (faster-whisper)
- Voice conversation falls back to whole-text playback (`playSpeechText`) — no streaming TTS backend

## Steps to reproduce

1. Start a voice conversation (mic → hands-free session).
2. Say something; wait for the reply to be spoken fully (status goes back to *listening*).
3. Say something else and go quiet so the turn is submitted by silence detection (do **not** interrupt during playback).
4. While the model is generating, the reply from step 2 is spoken again ("Speaking response" indicator). It stops when the new reply arrives.

## Evidence

Gateway log for the second turn — TTS is generated **before** the new prompt is even accepted, and the audio has exactly the same size as the previous turn's reply (38,016 bytes):

```
17:19:31,965 INFO tools.transcription_tools: Transcribed hermes-desktop-voice-oxjd2048.webm via local whisper (base, lang=ja, 6.7s audio)
17:19:32,140 INFO tools.tts_tool: Generating speech with Edge TTS...
17:19:32,282 INFO tui_gateway.server: tui prompt accepted: ... kind=user chars=23
17:19:32,462 INFO tools.tts_tool: TTS audio saved: .../tts_20260829_171932_138877.mp3 (38,016 bytes, provider: edge)   <- previous reply
17:19:34,933 INFO agent.conversation_loop: Turn ended: reason=text_response
17:19:35,409 INFO tools.tts_tool: TTS audio saved: .../tts_20260829_171935_025804.mp3 (27,792 bytes, provider: edge)   <- new reply
```

(Previous turn: `tts_20260829_171917_098279.mp3 (38,016 bytes)`.)

## Root cause

`apps/desktop/src/app/chat/composer/hooks/use-voice-conversation.ts`, `handleTurn` (the normal, silence-detected submit path):

```ts
awaitingSpokenResponseRef.current = true
dropSpeechSession()
await onSubmit(transcript)
setStatus('thinking')
```

Unlike the barge-in submit path (`submitCapturedUtterance`), it never calls `consumePendingResponse()` before submitting. Once `awaitingSpokenResponseRef` is set, the turn-drive effect evaluates `pendingResponse()`; whenever the previous reply is still considered unspoken at that moment, `openLiveSpeech(previousReply.id)` runs, streaming is unavailable, and `awaitFallbackSpeech` plays the stale text through `playSpeechText` until the new reply's id takes over.

## Fix

Mark the current reply spoken before submitting, exactly like the barge path does (one-line change + regression test). PR to follow.

Related but different: #91391 addresses duplicate auto-speak replays in `use-composer-voice.ts`; this issue is in the conversation loop's submit path.

tk-pkm111:
## What does this PR do?

Stops the Desktop voice conversation loop from re-speaking the **previous** assistant reply while the model is generating the next one.

After a reply finished playing, submitting the next voice turn via the normal (silence-detected) path set `awaitingSpokenResponseRef` and submitted — but never marked the just-played reply as spoken. The turn-drive effect then saw `awaiting` + an "unspoken" previous reply, opened speech for it, fell back to whole-text playback (`playSpeechText`), and the old answer played until the new reply arrived and barged in. The barge-in submit path (`submitCapturedUtterance`) already calls `consumePendingResponse()` before `onSubmit`; this PR gives the normal path the same treatment. A submitted turn makes any existing reply stale by definition, so consuming it before submit is the right invariant regardless of how the spoken anchor got out of sync.

## Related Issue

Fixes #97788

## Type of Change

## Changes Made

- `apps/desktop/src/app/chat/composer/hooks/use-voice-conversation.ts`: call `consumePendingResponse()` before `onSubmit(transcript)` in `handleTurn` (barge-path parity) and add it to the callback deps.
- `apps/desktop/src/app/chat/composer/hooks/use-voice-conversation-stale-reply.test.tsx`: regression test — with a previous reply still reported as unspoken, submitting the next turn must consume it **before** `onSubmit` and must never hand it to `playSpeechText`. Fails on `main` (`expected [ 'submit' ] to deeply equal [ 'consume', 'submit' ]`), passes with the fix.

## How to Test

1. Desktop, `tts.provider: edge` (any provider without streaming TTS works), start a voice conversation.
2. Say something, let the reply play fully.
3. Say something else and go quiet (no interruption). On `main`, the previous reply is spoken again while the model thinks; with this PR it stays silent until the new reply.
4. `cd apps/desktop && ../../node_modules/.bin/vitest run src/app/chat/composer/hooks/` → 107 tests pass (106 existing + 1 new). `tsc --noEmit -p tsconfig.json` clean.

Gateway log on `main` for step 3 (TTS generated before the new prompt is accepted, same byte size as the previous reply):

```
17:19:31,965 Transcribed hermes-desktop-voice-oxjd2048.webm via local whisper
17:19:32,140 Generating speech with Edge TTS...
17:19:32,282 tui prompt accepted: ... kind=user chars=23
17:19:32,462 TTS audio saved: tts_20260829_171932_138877.mp3 (38,016 bytes)   <- previous reply (same size as tts_20260829_171917_098279.mp3)
17:19:35,409 TTS audio saved: tts_20260829_171935_025804.mp3 (27,792 bytes)   <- new reply
```

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate (#91391 / #92334 touch adjacent code but not this path)
- [x] I've run `pytest tests/ -q` and all tests pass — N/A, TypeScript-only change; desktop vitest suite for the composer hooks passes
- [x] I've tested on my platform: macOS 15 (Apple Silicon), Hermes Desktop

- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — renderer-only logic, no platform-specific code

### DIFF
--- a/apps/desktop/src/app/chat/composer/hooks/use-voice-conversation-stale-reply.test.tsx
+++ b/apps/desktop/src/app/chat/composer/hooks/use-voice-conversation-stale-reply.test.tsx
@@ -1,0 +1,138 @@
+import { act, cleanup, renderHook, waitFor } from '@testing-library/react'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import type { MicRecording } from './use-mic-recorder'
+import { useVoiceConversation } from './use-voice-conversation'
+
+// Regression: after a reply finished playing, the NEXT voice turn must not
+// re-speak that reply while the model is thinking. The normal (silence-detected)
+// submit path has to mark the previous reply spoken BEFORE submitting — the
+// barge path already did — otherwise the turn-drive effect sees `awaiting` plus
+// an "unspoken" previous reply and plays it through the whole-text fallback
+// until the new answer arrives and barges in.
+
+const playSpeechText = vi.fn(async (_text: string, _options: unknown) => true)
+
+vi.mock('@/lib/voice-barge-in', () => ({
+  monitorSpeechDuringPlayback: () => vi.fn()
+}))
+
+vi.mock('@/lib/voice-playback', () => ({
+  markVoicePlaybackInterrupted: vi.fn(),
+  playSpeechText: (text: string, options: unknown) => playSpeechText(text, options),
+  startSpeechStream: vi.fn(async () => null),
+  stopVoicePlayback: vi.fn()
+}))
+
+vi.mock('@/lib/thinking-sound', () => ({
+  startThinkingSound: vi.fn(),
+  stopThinkingSound: vi.fn()
+}))
+
+const micHandle = {
+  cancel: vi.fn(),
+  start: vi.fn(async () => undefined),
+  stop: vi.fn<() => Promise<MicRecording | null>>(async () => null)
+}
+
+vi.mock('./use-mic-recorder', () => ({
+  useMicRecorder: () => ({ handle: micHandle, level: 0, recording: false })
+}))
+
+vi.mock('@/i18n', () => ({
+  useI18n: () => ({
+    t: {
+      notifications: {
+        voice: {
+          configureSpeechToText: 'configure STT',
+          couldNotStartSession: 'could not start',
+          microphoneFailed: 'mic failed',
+          playbackFailed: 'playback failed',
+          transcriptionFailed: 'transcription failed',
+          unavailable: 'unavailable'
+        }
+      }
+    }
+  })
+}))
+
+vi.mock('@/store/notifications', () => ({
+  notify: vi.fn(),
+  notifyError: vi.fn()
+}))
+
+describe('useVoiceConversation stale reply on next turn', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    micHandle.start.mockResolvedValue(undefined)
+    micHandle.stop.mockResolvedValue(null)
+  })
+
+  afterEach(cleanup)
+
+  it('marks the previous reply spoken before submitting so it is not re-spoken while thinking', async () => {
+    const calls: string[] = []
+
+    // The previous turn's answer still sits at the bottom of the thread and —
+    // as in the field — is NOT yet marked spoken when the user's next
+    // utterance lands. It stays "pending to speak" until consumed.
+    let staleReplyUnspoken = true
+
+    const consumePendingResponse = vi.fn(() => {
+      calls.push('consume')
+      staleReplyUnspoken = false
+    })
+
+    // Mirrors the real app: submitting a turn makes the agent busy.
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed `apps/desktop/src/app/chat/composer/hooks/use-voice-conversation.ts` + a new test for PR #97789. The fix is exactly the right shape: the silence-detected submit path now calls `consumePendingResponse()` *before* `await onSubmit(transcript)`, matching the barge path, so the just-played reply is marked spoken before the next turn drives the "unspoken previous reply" whole-text fallback. The test is well-built — it models the stale-unspoken reply, asserts the `consume`→`submit` ordering, and verifies `playSpeechText` is never called while thinking.

Non-blocking observations:

1. **`consumePendingResponse` is new in the `useCallback` deps array.** If `consumePendingResponse` is not itself memoized upstream (stable identity across renders), the `stopTurn` callback will now be recreated on every render, which can re-trigger any effect that depends on it. If it's already `useCallback`-stable this is a no-op — worth confirming upstream to avoid a subtle re-render/re-subscribe churn in the hook graph.
2. **`consumePendingResponse()` runs before `await onSubmit(...)`, so a failed submit still consumes the reply.** Since the reply was already played (the whole premise of "stale"), marking it spoken even when the new turn fails to submit is the correct trade-off — it only means a genuinely unplayed reply (e.g. the fallback path) could be skipped. Given the comment explicitly scopes this to the just-finished-playback case, this is fine; just confirming the semantics are intentional for the submit-failure edge.
3. **Minor:** the `pendingResponse()` callback in the test returns a "stale" reply whenever `staleReplyUnspoken` is true, which is a faithful model of the field condition described in the PR body. Good coverage of the actual failure mode rather than a mock-shaped one.

Correct, minimal, and well-tested; nothing blocking.

PR #97798, opened 2026-08-29
fix(update): surface why a bundled-skill sync failed instead of a bare "sync failed"

### LINKED ISSUE #97792: update: bundled-skill sync prints bare 'sync failed' — child stderr discarded, in-process exception only at debug level
## Bug Description

When `hermes update` syncs bundled skills into profiles, a failure prints the bare string `sync failed` with **no diagnostic information whatsoever** — the child process's stderr is captured and discarded, and the in-process failure above it goes to `logger.debug`, which does not reach `~/.hermes/logs/errors.log`. There is no way to find out what went wrong after the fact.

Real output from a v0.20.5 → v0.20.6 update on a 7-profile install:

```
→ Syncing bundled skills...

→ Syncing bundled skills to all profiles...
  default: sync failed
  creative: ~3 user-modified
  design: sync failed
  dev: sync failed
  finance: sync failed
  general: sync failed
  research: sync failed
```

Note also the first block (`→ Syncing bundled skills...`) printed **no result line at all** — that is the in-process `sync_skills()` having thrown, swallowed at `hermes_cli/update_cmd.py:8645-8646`.

Re-running the identical code path by hand afterwards succeeded for all 7 profiles in ~0.06s each, so the failure was transient — and completely unattributable.

## Where it goes wrong

**1. Child stderr is discarded when `quiet=True`** — `hermes_cli/profiles.py:1402-1416`:

```python
if result.returncode == 0 and result.stdout.strip():
    return json.loads(result.stdout.strip())
if not quiet:                                     # <- update_cmd calls with quiet=True
    print(f"⚠ Skill seeding returned exit code {result.returncode}")
    if result.stderr.strip():
        print(f"  {result.stderr.strip()[:200]}")
return None
```

The caller passes `quiet=True` (`update_cmd.py:8665`), so `returncode` and `stderr` are both dropped on the floor. `None` comes back carrying no information, and becomes the string `sync failed` at `update_cmd.py:8681`.

Note `json.loads()` on line 1403 is also unguarded — a malformed stdout raises `JSONDecodeError`, which is *not* in the `except` list (`TimeoutExpired`, `Exception` is caught at 1413 so it is swallowed there instead, again losing the payload).

**2. The in-process failure is debug-only** — `hermes_cli/update_cmd.py:8645-8646`:

```python
except Exception as e:
    logger.debug("Skills sync during update failed: %s", e)
```

At default log level this is invisible, and nothing lands in `errors.log`. Confirmed: grepping `errors.log` around the update timestamp shows unrelated warnings only, nothing about skills.

## Expected Behavior

A failure during a step the updater explicitly reports on should say *why*, at minimum:

- the child's exit code
- the first N lines of its stderr
- the exception type/message for the in-process path, at `warning` level so it reaches `errors.log`

## Actual Behavior

`sync failed` — and nothing else, anywhere on disk.

## Suggested fix

Report the reason regardless of `quiet` (quiet should suppress *success* chatter, not error detail). Something like:

```python
# profiles.py — return the reason instead of a bare None
if result.returncode != 0 or not result.stdout.strip():
    reason = (result.stderr.strip().splitlines() or ["no stderr"])[-1][:200]
    return {"ok": False, "error": f"rc={result.returncode}: {reason}"}
```

and let `update_cmd.py` print `f"  {p.name}: sync failed ({r['error']})"`. Also promote the in-process `logger.debug` at 8646 to `logger.warning` (or `exc_info=True` at debug plus a one-line warning) so a swallowed exception is recoverable from `errors.log`.

## Environment

- Hermes v0.20.6, `main @ aff5125f8e` (update from v0.20.5)
- 7 profiles; skills symlinked into a shared `~/.hermes/skill-store/`
- Timeout ruled out as the cause: each profile's sync takes ~0.06s against the 60s budget

liuhao1024:
## What does this PR do?

When `hermes update` syncs bundled skills into profiles and a profile's sync fails, the user sees the bare string `sync failed` with no diagnostic information anywhere — the child process's exit code and stderr are captured and then discarded, and the in-process sync failure is logged only at debug level, so it never reaches `~/.hermes/logs/errors.log` either (#97792).

Root cause: `seed_profile_skills()` in `hermes_cli/profiles.py` returns `None` for **every** failure mode (non-zero exit, empty stdout, timeout, exception) whenever the caller passes `quiet=True`, dropping the return code and stderr on the floor. The update loop in `hermes_cli/update_cmd.py` then has nothing to print but "sync failed".

The fix follows the shape suggested in the issue:

- `seed_profile_skills()` failure paths now return an `{"error": ...}` dict describing the failure — `rc=<code>: <last stderr line>` for child failures, `"timed out after 60s"` for timeouts, `"<ExceptionType>: <msg>"` for everything else (including a malformed child stdout). `quiet` now only suppresses success chatter, not error detail. The non-quiet `print` behavior is unchanged.
- `hermes_cli/update_cmd.py` prints `sync failed (rc=1: ValueError: bad manifest)` instead of a bare `sync failed`, and promotes the in-process sync exception from `logger.debug` to `logger.warning` so it lands in `errors.log`.
- `hermes_cli/main.py` (the `hermes profile create` seeding path, the other caller that consumes the return value) prints the reason in its "Skills could not be seeded" message. The other two callers (`web_routers/profiles.py`, `tui_gateway/methods_profiles.py`) ignore the return value and are unaffected.

One subtlety worth noting: without the new `elif r.get("error")` branches, a returned error dict is truthy and would be misreported as a *successful* sync with zero copies ("up to date"), so the callers had to learn about the new shape in the same change. The tests pin this: a failing profile must print `sync failed (rc=...)`, never `up to date`.

## Related Issue

Fixes #97792

## Type of Change

## Changes Made

- `hermes_cli/profiles.py` — `seed_profile_skills()` failure paths return `{"error": ...}` (rc + last stderr line / timeout / exception) instead of `None`; docstring updated
- `hermes_cli/update_cmd.py` — profile sync loop prints the error reason (`sync failed (rc=...)`); in-process sync exception logged at `warning` instead of `debug`
- `hermes_cli/main.py` — profile-create seeding path surfaces the error reason in the retry hint
- `tests/hermes_cli/test_profiles.py` — 4 new tests pinning the failure-return contract: non-zero exit with stderr tail, rc=0/empty stdout, timeout, malformed stdout (JSONDecodeError)
- `tests/hermes_cli/test_cmd_update.py` — new test pinning that a failed profile prints `sync failed (rc=1: ...)` and is never masked as `up to date`

## How to Test

1. `pytest tests/hermes_cli/test_profiles.py -q` — Observed result: 60 passed, 2 skipped (4 of them the new `TestSeedProfileSkillsFailureReason` cases).
2. `pytest tests/hermes_cli/test_cmd_update.py -q -k CmdUpdateProfileSkillSync` — Observed result: 3 passed, including the new `test_failed_profile_sync_prints_reason`.
3. Manual reproduction of the issue shape: make a seeded profile's child sync fail (e.g. point the subprocess at a broken state) and run `hermes update` — Observed result before: `design: sync failed` and nothing in any log; after: `design: sync failed (rc=1: <last stderr line>)` and a `WARNING Skills sync during update failed: ...` line when the in-process path fails.
4. `ruff check` on the five touched files — Observed result: all checks passed. (`ruff format --check` reports the same pre-existing reformat list on these files as on a clean `upstream/main` checkout; no new formatting debt is introduced.)

Note on the local run environment for step 2: on a dev machine that runs real `ai.hermes.gateway*` LaunchAgents, the cmd_update integration tests hit the live fleet-restart gate (`sys.exit(1)` at update_cmd.py:9984) before reaching their assertions — the same happens on a clean `upstream/main` checkout, so it is a pre-existing local-environment artifact, not caused by this diff. CI's Linux runners have no launchd and do not take that path.

### Code

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (docstring of `seed_profile_skills` updated)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (no platform-specific code touched)

## For New Skills

N/A — no new skill.

## Screenshots / Logs

Before (from the issue, v0.20.5 → v0.20.6 update on a 7-profile install):

```
→ Syncing bundled skills to all profiles...
  default: sync failed
  creative: ~3 user-modified
  design: sync failed
  ...
```

After (pinned by `test_failed_profile_sync_prints_reason`):

```
→ Syncing bundled skills to all profiles...
  default: up to date
  design: sync failed (rc=1: ValueError: bad manifest)
```

### DIFF
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -11134,6 +11134,11 @@ def cmd_profile(args):
                         "No bundled skills seeded (--no-skills). "
                         "Delete .no-bundled-skills in the profile to opt back in."
                     )
+                elif result and result.get("error"):
+                    print(
+                        "⚠ Skills could not be seeded ({}). "
+                        "Run `{} update` to retry.".format(result["error"], name)
+                    )
                 elif result:
                     copied = len(result.get("copied", []))
                     print(f"{copied} bundled skills synced.")
--- a/hermes_cli/profiles.py
+++ b/hermes_cli/profiles.py
@@ -1381,7 +1381,9 @@ def seed_profile_skills(profile_dir: Path, quiet: bool = False) -> Optional[dict
     """Seed bundled skills into a profile via subprocess.
 
     Uses subprocess because sync_skills() caches HERMES_HOME at module level.
-    Returns the sync result dict, or None on failure.
+    Returns the sync result dict, or a ``{"error": ...}`` dict describing the
+    failure (exit code + last stderr line, timeout, or exception) — callers
+    can surface the reason even when ``quiet=True`` (#97792).
 
     Profiles that opted out of bundled skills (via ``hermes profile create
     --no-skills`` — which writes ``.no-bundled-skills`` to the profile root)
@@ -1401,19 +1403,22 @@ def seed_profile_skills(profile_dir: Path, quiet: bool = False) -> Optional[dict
         )
         if result.returncode == 0 and result.stdout.strip():
             return json.loads(result.stdout.strip())
+        # quiet suppresses success chatter, not error detail: keep returning
+        # the reason so `hermes update` can print why a profile failed (#97792).
+        stderr_tail = result.stderr.strip().splitlines() or ["no stderr"]
         if not quiet:
             print(f"⚠ Skill seeding returned exit code {result.returncode}")
             if result.stderr.strip():
                 print(f"  {result.stderr.strip()[:200]}")
-        return None
+        return {"error": f"rc={result.returncode}: {stderr_tail[-1][:200]}"}
     except subprocess.TimeoutExpired:
         if not quiet:
             print("⚠ Skill seeding timed out (60s)")
-        return None
+        return {"error": "timed out after 60s"}
     except Exception as e:
         if not quiet:
             print(f"⚠ Skill seeding failed: {e}")
-        return None
+        return {"error": f"{type(e).__name__}: {e}"}
 
 
 def backfill_profile_envs(quiet: bool = False) -> List[str]:
--- a/hermes_cli/profiles.py
+++ b/hermes_cli/profiles.py
@@ -1377,6 +1377,18 @@ def create_profile(
     return profile_dir
 
 
+# The failure reason flows into user-facing status lines (`hermes update`,
+# `hermes profile create`); a raw stderr tail from a crashing child can carry
+# ANSI escape sequences or carriage resets that garble those lines (#97792).
+_REASON_ESCAPE_RE = re.compile(r"\x1b\[[0-9;?]*[A-Za-z]|\x1b\][^\x07]*\x07")
+
+
+def _reason_text(stderr: str) -> str:
+    # Carriage returns become spaces *before* any splitlines() so a mid-line
+    # \r cannot turn the tail into an empty last line.
+    return _REASON_ESCAPE_RE.sub("", stderr).replace("\r", " ")
+
+
 def seed_profile_skills(profile_dir: Path, quiet: bool = False) -> Optional[dict]:
     """Seed bundled skills into a profile via subprocess.
 
@@ -1405,12 +1417,12 @@ def seed_profile_skills(profile_dir: Path, quiet: bool = False) -> Optional[dict
             return json.loads(result.stdout.strip())
         # quiet suppresses success chatter, not error detail: keep returning
         # the reason so `hermes update` can print why a profile failed (#97792).
-        stderr_tail = result.stderr.strip().splitlines() or ["no stderr"]
+        stderr_tail = _reason_text(result.stderr).strip().splitlines() or ["no stderr"]
         if not quiet:
             print(f"⚠ Skill seeding returned exit code {result.returncode}")
             if result.stderr.strip():
                 print(f"  {result.stderr.strip()[:200]}")
-        return {"error": f"rc={result.returncode}: {stderr_tail[-1][:200]}"}
+        return {"error": f"rc={result.returncode}: {stderr_tail[-1].strip()[:200]}"}
     except subprocess.TimeoutExpired:
         if not quiet:
             print("⚠ Skill seeding timed out (60s)")
--- a/hermes_cli/update_cmd.py
+++ b/hermes_cli/update_cmd.py
@@ -8643,7 +8643,7 @@ def _cmd_update_impl(args, gateway_mode: bool):
             if not result["copied"] and not result.get("updated"):
                 print("  ✓ Skills are up to date")
         except Exception as e:
-            logger.debug("Skills sync during update failed: %s", e)
+            logger.warning("Skills sync during update failed: %s", e)
 
         # Sync bundled skills to all profiles (including the active one).
         # seed_profile_skills() uses subprocess with an explicit HERMES_HOME so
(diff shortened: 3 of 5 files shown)


Enough1122:

Reviewed `hermes_cli/profiles.py`, `hermes_cli/update_cmd.py`, `hermes_cli/main.py` + tests for PR #97798. The change correctly fixes #97792: `seed_profile_skills()` no longer collapses every failure to `None`, instead returning `{"error": ...}` (rc + stderr tail, timeout, or exception), and `hermes update` surfaces that reason in the per-profile status line. The `logger.debug` → `logger.warning` bump is also right so the in-process failure reaches `errors.log`. Tests cover nonzero exit, zero-exit-empty-stdout, timeout, malformed stdout, and the update-loop display.

Non-blocking observations:

1. **The fresh-create call site in `main.py` (~line 9455) still treats a dict as success.** `hermes profile create` does `if result and result.get("skipped_opt_out"): … elif result: print(f"{copied} bundled skills synced.") else: print("⚠ Skills could not be seeded…")`. With the contract change, a failed seed now returns a truthy `{"error": …}` dict, so this path hits `elif result:` and prints "0 bundled skills synced." instead of the failure warning. The PR updates the `cmd_profile`/update status lines but not this fresh-create caller — the `elif result:` should also check for `result.get("error")` (like the new branch at line ~11134). This is the one spot where the API change silently degrades a message.
2. **`rc=0` with empty stdout is reported as an error.** A zero exit with no JSON on stdout now yields `{"error": "rc=0: no stderr"}`. If a genuinely successful no-op sync (nothing to seed, child prints nothing) ever exits 0 with empty stdout, it would be displayed as `sync failed (rc=0: no stderr)` — a false positive. The test asserts this exact shape is treated as an error, so it's deliberate; just flagging that distinguishing "empty success" from "failure" (e.g. requiring the child to always emit a JSON result, or treating `rc==0` as success) would avoid a confusing message.
3. **`web_routers/profiles.py` calls with `quiet=True` and ignores the return value** — unaffected by the contract change (the dict is simply dropped). Fine.
4. **The error string embeds the raw stderr tail (capped at 200 chars).** Reasonable for a CLI status line, but the reason now flows into `update` output that may be shown to users — stderr can contain control characters or ANSI. A quick sanitize (strip escape sequences) would keep the message clean. Cosmetic.

The core fix is right and well-tested; the `main.py` fresh-create branch (#1) is worth aligning before merge.

liuhao1024:
Thanks for the thorough review. Walking the four observations:

1. **fresh-create caller** — already covered in this PR: `hermes_cli/main.py` has the `elif result and result.get("error")` branch (the "⚠ Skills could not be seeded (…)" arm) since fd528aeb, so a truthy error-dict no longer prints "0 bundled skills synced." there. The ~line-9455 reference predates that commit's diff context.
2. **rc=0 + empty stdout as error** — deliberate fail-closed: the child always prints a JSON result on success (even a no-op sync emits a result object), so an rc=0 with no stdout means the child died before reporting; treating that as success would silently hide a broken seed.
3. Acknowledged — the `quiet=True` call site dropping the dict is unchanged behavior.
4. **stderr control characters** — fair point, fixed in 922177f: the reason now strips ANSI escapes and turns lone carriage returns into spaces *before* splitting for the tail line (a mid-line \\r previously flattened the tail to an empty last line via `splitlines()`), with a regression test pinning the sanitized shape.

PR #97799, opened 2026-08-29
fix(gateway): scope /fast global persistence to profile

ricardosalta:
## Summary

In a multiplex gateway, open `/fast --global` from a named profile and choose
a mode after the command returns: the callback saves to the default profile's
config. Capture the originating profile and re-enter its runtime scope while
applying the delayed choice, following the existing `/model` picker pattern.

The fix uses the upstream profile-aware config writer and preserves all four
modes (`fast`, `normal`, `auto`, `cold`), session-only choices, and the session
fallback when a global write fails.

## Testing

- 25 focused tests pass through `scripts/run_tests.sh`: fast command,
  reasoning command, and model command profile config suites.
- Delayed-callback coverage runs after the originating scope exits and checks
  named/default config isolation, clearing existing overrides on global success,
  session-only choices, and failed-save fallback across all four modes.
- Restoring upstream callback behavior makes three global picker cases fail
  because the default profile's real temporary config is modified.
- Ruff and `git diff --check` pass.

## Scope

Focused gateway bug fix; no new configuration or dependencies. Validated on Linux
with real temporary config file reads/writes. No live messaging account required.


Enough1122:

Reviewed `gateway/slash_commands.py` + tests for PR #97799. This is the `/fast` sibling of the `/reasoning` global-scope fix (PR #97775): `_command_profile_home` is resolved (only when `persist_global` AND multiplex is active — tighter than the `/reasoning` variant's unconditional multiplex resolution) and threaded into `_save_gateway_config_key` so a typed `/fast fast --global` and a delayed `/fast --global` picker tap both persist to the originating multiplex profile. `_apply_fast_selection` is a nested closure over `_command_profile_home`, so both typed and picker paths are covered. The two tests are strong: typed-path isolation (default untouched, named profile updated) and delayed-picker isolation (tap fires after `_profile_runtime_scope` exits and still lands in the named profile).

Non-blocking observations:

1. **Merge-conflict risk with PR #97775 — same files, same hunks.** Both PRs modify `_save_gateway_config_key` with the identical `config_home` signature and adjacent regions of `gateway/slash_commands.py` (same `@@ -3645,12 +3645,14 @@` context). They're sibling fixes by the same author, but each diff is based on the same pre-merge `main`; whichever merges second will conflict with the first. Worth rebasing one on the other before merge so the shared `_save_gateway_config_key` signature and the `_profile_runtime_scope` pattern land once.
2. **Inconsistent profile-home resolution conditions between the two PRs.** 97775 resolves `_command_profile_home` on `multiplex_profiles` alone; 97799 requires `persist_global` too. The `persist_global` guard here is arguably more correct (no need to resolve a profile home for a non-persisting picker), but the asymmetry means the two handlers resolve profile scope differently. Consider aligning them (or documenting why `/reasoning` resolves unconditionally).
3. **Same deferred-scope caveat as `/reasoning`:** on a delayed tap, `_set_session_service_tier_override(session_key, None, clear=True)` clears the override keyed by the *command-time* `session_key` (closure-captured, so it's the originating profile's key — correct), while the config write goes to `_command_profile_home`. Consistent and correct; just noting both PRs share the pattern and its test coverage is good.
4. **Minor:** the picker test re-enters `_profile_runtime_scope(named_home)` around the override assertion; since `_session_service_tier_overrides` is process-shared the scope isn't strictly needed for the assertion, but it mirrors the production shape — fine.

Correct, well-tested, and consistent with the sibling fix; nothing blocking beyond the merge-ordering note.

ricardosalta:
⌾ Refreshed onto current upstream `main` (`b7d7d2929a`) without force-pushing. Current head: `796dd4f5f0`.

Relevance recheck: delayed `/fast --global` picker persistence can still escape the originating multiplex profile on current `main`; upstream has not superseded this fix. The production patch still applies cleanly.

One regression-test seam had moved upstream from the old generic adapter lookup to `_delivery_adapter_for`; the test now follows that current production seam. Validation on the refreshed head: 25 focused gateway tests passed (`test_fast_command.py`, `test_reasoning_command.py`, `test_model_command_profile_config.py`), Ruff passed, and `git diff --check` is clean.

PR #56504, opened 2026-07-01
fix: strip anthropic_content_blocks in chat_completions transport

KenKozuma:
## Summary

Fixes HTTP 400 `Extra inputs are not permitted: messages[N].anthropic_content_blocks` when switching from an `anthropic_messages`-routed model (e.g. MiniMax on opencode-go) to a `chat_completions`-routed model (e.g. GLM-5.2, DeepSeek V4 Pro) mid-session.

## Problem

```bash
# Start with minimax-m3 (anthropic_messages)
$ hermes model
# → minimax-m3 works fine, generates thinking blocks as anthropic_content_blocks

# Switch to GLM-5.2-med (chat_completions) mid-session
$ /model glm-5.2-med

# Next API call fails:
API call failed after 3 retries: HTTP 400:
Extra inputs are not permitted, field: 'messages[146].anthropic_content_blocks'
```

The `anthropic_messages` transport writes `anthropic_content_blocks` into assistant messages to preserve thinking-block ordering for replay. When the model switches to `chat_completions`, these fields leak into the outgoing payload. Strict providers (opencode-go, Fireworks) reject unknown top-level message keys.

## Root Cause

`agent/transports/chat_completions.py` — `convert_messages()` already strips several provider-specific fields (`codex_reasoning_items`, `codex_message_items`, `tool_name`, `timestamp`) but did not handle `anthropic_content_blocks`.

The existing `test_anthropic_output_field_leak.py` only covers block-level field sanitization (`parsed_output`, `caller`), not the transport-level message key stripping.

## Fix

`agent/transports/chat_completions.py` — added `anthropic_content_blocks` to both the detection check and the stripping loop in `convert_messages()` (+4 lines):

```python
# Detection (line 183)
or "anthropic_content_blocks" in msg  # model-switch leak

# Stripping (line 214)
msg.pop("anthropic_content_blocks", None)  # model-switch leak
```

The `_anthropic_content_blocks` variant (stashed by `model_metadata.py` for image counting) is already handled by the existing `_`-prefixed key stripping logic (line 209).

## Related

- **PR #56216** — fixes OpenCode base_url `/v1` missing bug. Both bugs surface when switching models on opencode-go, but have different root causes and disjoint fix locations.

## Testing

13 tests across 2 files, all passing:

```
test_anthropic_output_field_leak.py (existing, 7 tests):
  ✓ test_text_block_strips_parsed_output_and_null_citations
  ✓ test_tool_use_strips_caller
  ✓ test_thinking_preserves_signature
[...]

PR #97800, opened 2026-08-29
fix(agent): strip anthropic_content_blocks on the chat_completions wire

ZoloZiak:
## What

`ChatCompletionsTransport.convert_messages` did not strip the
`anthropic_content_blocks` message field before sending a request on the
OpenAI (chat_completions) wire. Add it to both the sanitize-detection
predicate and the strip block, alongside the existing
`codex_reasoning_items` / `codex_message_items` / `timestamp` /
`api_content` internal-field scrubbing.

## Why

`anthropic_content_blocks` is a Hermes-internal, Anthropic-only channel. It
is attached to an assistant turn by `normalize_response` **only** when that
turn interleaves *signed thinking* blocks with `tool_use`, so that the
Anthropic adapter can replay the turn verbatim and keep the thinking-block
signatures valid (reconstructing from the parallel `reasoning_details` +
`tool_calls` lists reorders the blocks and invalidates the signatures).

On the native Anthropic Messages wire the adapter consumes this field, so it
never reaches the API. But nothing stripped it on the **chat_completions**
path. As soon as such a turn is replayed through a chat_completions target —
an OpenAI-compatible proxy, or a fallback from an Anthropic primary to an
OpenAI-wire provider — the field leaks onto the wire and a strict provider
rejects the whole request:

```
HTTP 400: 'messages.N' : for 'role:assistant' the following must be
satisfied[('messages.N' : property 'anthropic_content_blocks' is unsupported)]
```

Because the offending turn sits in history, every subsequent request in the
session hits the same 400 until the turn is compacted away.

This is the same bug class as the `codex_reasoning_items` / `timestamp` /
`api_content` leaks that are already handled a few lines above — an
internal-only message field escaping onto a strict OpenAI-compatible wire.

## Scope

`reasoning_details` is intentionally **not** stripped here: it has its own
provider-aware stripper and is legitimately required on the OpenRouter wire.
This change only removes the Anthropic-only replay channel, which is never
valid on any chat_completions request.

## Reproduction

Build an assistant message carrying `anthropic_content_blocks` (a signed
`thinking` block followed by a `tool_use` block) plus the parallel
`tool_calls`, and run it through `convert_messages(model=<any OpenAI-wire
model>)`. Before: the outgoing message still contains
`anthropic_content_blocks`. After: it is stripped, leaving `content`,
`role`, and `tool_calls`.

## Test

Added coverage asserts that `convert_messages` drops
`anthropic_content_blocks` from an assistant turn while preserving
`content`, `role`, and `tool_calls`.

### DIFF
--- a/agent/transports/chat_completions.py
+++ b/agent/transports/chat_completions.py
@@ -331,6 +331,13 @@ class ChatCompletionsTransport(ProviderTransport):
                 or "effect_disposition" in msg
                 or "timestamp" in msg  # #47868 — strict providers reject this
                 or "api_content" in msg  # persist-what-you-send sidecar
+                # Anthropic-only interned replay channel (set by
+                # normalize_response for interleaved signed-thinking + tool_use
+                # turns). Consumed by the anthropic adapter; NEVER valid on an
+                # OpenAI-wire request, so a chat_completions target (Palantir
+                # GPT/Gemini proxy, Groq, …) rejects it with HTTP 400
+                # ("property 'anthropic_content_blocks' is unsupported").
+                or "anthropic_content_blocks" in msg
             ):
                 needs_sanitize = True
                 break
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #56504: both strip anthropic_content_blocks from ChatCompletionsTransport replay messages before strict OpenAI-compatible targets receive them.

ZoloZiak:
Closing as a duplicate of #56504, which is the earlier open PR making the same
fix (stripping `anthropic_content_blocks` in `ChatCompletionsTransport` before
strict OpenAI-compatible targets receive it). I missed the existing PR in my
search before opening this — apologies for the noise. Happy to help review
#56504 if that's useful.

PR #84073, opened 2026-08-11
fix(tests): make Windows pytest collection portable

omerbek:
## Summary
- avoid calling POSIX-only `os.geteuid()` while pytest is importing the doctor journal-mode tests on Windows
- use `tempfile.gettempdir()` for the parallel-runner handoff directory instead of assuming `/tmp`
- make the nested database path assertion use the platform path separator

Fixes #83935.

## Testing
- `.venv\Scripts\python.exe -m pytest --collect-only -q tests\hermes_cli\test_doctor_journal_modes.py tests\test_run_tests_parallel.py`
- `.venv\Scripts\python.exe -m pytest -q tests\hermes_cli\test_doctor_journal_modes.py`

Note: running `tests\test_run_tests_parallel.py` fully on this Windows checkout still times out inside its subprocess runner smoke test after collection succeeds; this PR targets the collection-time abort described in the issue.

PR #74083, opened 2026-07-29
fix(tests): use the platform temp dir for the isolation-probe handoff

hyqqx:
## What does this PR do?

`tests/test_run_tests_parallel.py` builds its handoff directory from
`os.environ.get("TMPDIR", "/tmp")` and `mkdir()`s it at **module scope**. On Windows
`TMPDIR` is normally unset — the platform uses `TEMP`/`TMP` — so the fallback resolves to a
non-existent `\tmp` on the current drive and raises during collection:

```
tests\test_run_tests_parallel.py:38: in <module>
    _HANDOFF_DIR.mkdir(exist_ok=True)
E   FileNotFoundError: [WinError 3] The system cannot find the path specified:
    '\tmp\hermes-isolation-probe'
```

pytest aborts the whole run on a collection error, so this single line stops the entire
suite on native Windows.

## The module already declares the right intent

Its docstring says *"POSIX-only … Marked accordingly"* and the probe carries
`@pytest.mark.skipif(sys.platform == "win32", reason="POSIX-only probe")`. The marker never
gets a chance to apply, because the `mkdir` runs at import time — before pytest evaluates
any marker.

`tempfile.gettempdir()` is the cross-platform primitive for this, and it still honours
`$TMPDIR` first, so POSIX behaviour is byte-identical. Nothing else changes: `_HANDOFF_DIR`
is read only by `_handoff_path_for`, whose sole caller is the skipped POSIX probe.

## Scope change since the first review

This PR originally carried two collection-time fixes. The `import pty` half **landed
independently on `main`** as `05504bd9f` (*test(gateway): make test_gateway collectable on
Windows*, @iso2kx, 2026-07-30) while this was open — same approach, same call site. That
half is dropped here and the branch is **rebased onto current `main`**, addressing
@monerostar's note that the head was ~600 commits behind. What remains is the one blocker
still live on `main` today.

## How to test

From **PowerShell or cmd** (not Git Bash — MSYS sets `TMPDIR`, which masks this entirely):

```powershell
python -m pytest tests/test_run_tests_parallel.py --collect-only -q
```

**Before**, on current `main` (`710b02663`):
```
E   FileNotFoundError: [WinError 3] ... '\tmp\hermes-isolation-probe'
Interrupted: 1 error during collection
no tests collected, 1 error
```

**After:**
```
7 tests collected in 0.12s        # exit 0
```

Running the file: `3 passed, 1 skipped, 3 failed`. The three failures are pre-existing
[...]

PR #97801, opened 2026-08-29
fix(tests): stop pytest collection aborting on Windows

### LINKED ISSUE #83935: [Bug]: Windows: plain pytest aborts during collection, so the whole suite cannot run
## Bug Description

On Windows, running `pytest` over the whole suite aborts during **collection**, so zero tests execute. Two test modules do POSIX-only work at import time.

`scripts/run_tests_parallel.py` is unaffected — it spawns one pytest per file, so only those two files fail and the rest proceed. But anyone who runs plain `pytest` (the obvious thing to try, and what `pyproject.toml`'s `[tool.pytest.ini_options]` is configured for) gets nothing at all.

## Steps to Reproduce

```
> python -m pytest -q
```

## Expected Behavior

The suite runs. Platform-specific tests skip themselves.

## Actual Behavior

```
=================================== ERRORS ====================================
_______ ERROR collecting tests/hermes_cli/test_doctor_journal_modes.py ________
tests\hermes_cli\test_doctor_journal_modes.py:117: in TestReadJournalMode
    @pytest.mark.skipif(os.geteuid() == 0, reason="root ignores file permissions")
E   AttributeError: module 'os' has no attribute 'geteuid'
______________ ERROR collecting tests/test_run_tests_parallel.py ______________
tests\test_run_tests_parallel.py:38: in <module>
    _HANDOFF_DIR.mkdir(exist_ok=True)
E   FileNotFoundError: [WinError 3] ...: '\\tmp\\hermes-isolation-probe'
!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!
14 skipped, 86 deselected, 7 warnings, 2 errors in 121.22s
```

### 1. `os.geteuid` is POSIX-only

[tests/hermes_cli/test_doctor_journal_modes.py:117](https://github.com/NousResearch/hermes-agent/blob/main/tests/hermes_cli/test_doctor_journal_modes.py#L117) calls it inside a `skipif` decorator, which is evaluated at class-body execution time — so the module cannot even be imported on Windows. Guarding it inside the condition does not help for the same reason; the call has to be deferred:

```python
@pytest.mark.skipif(
    getattr(os, "geteuid", lambda: 1)() == 0,
    reason="root ignores file permissions",
)
```

### 2. `/tmp` is hardcoded at module scope

[tests/test_run_tests_parallel.py:38](https://github.com/NousResearch/hermes-agent/blob/main/tests/test_run_tests_parallel.py#L38) does `_HANDOFF_DIR.mkdir(exist_ok=True)` on a `/tmp`-rooted path at import time. On Windows that resolves to `\tmp\...` on the current drive, which does not exist. `tempfile.gettempdir()` would work everywhere.

## Suggested fix

Both are one-liners. Optionally add a Windows job to CI, or at least a smoke check that `pytest --collect-only` exits 0 on Windows — collection-time breakage is invisible to the per-file parallel runner that CI uses.

## Environment

- Hermes Agent v0.20.0 (2026.8.3), install method: git
- upstream base `c0106e50e`
- Python 3.11.15, Windows-10-10.0.26200-SP0

iamiskender:
Closes #83935.

On Windows, `python -m pytest` over the whole suite aborts during collection, so zero tests run. Two modules do POSIX-only work at import time. `scripts/run_tests_parallel.py` is unaffected because it spawns one pytest per file, which is why CI stays green while the suite is unrunnable locally.

## Measured, on Windows 11 / Python 3.11.15

| file | before | after |
|---|---|---|
| `tests/test_run_tests_parallel.py` | 0 collected, `FileNotFoundError: [WinError 3] '\tmp\hermes-isolation-probe'` | 10 collected |
| `tests/hermes_cli/test_doctor_journal_modes.py` | 0 collected, `AttributeError: module 'os' has no attribute 'geteuid'` | 36 collected |

Both re-run on WSL after the change: 46 collected, no errors.

## 1. `/tmp` hardcoded at module scope

`_HANDOFF_DIR` came from `os.environ.get("TMPDIR", "/tmp")`. `TMPDIR` is not set on Windows, so the fallback resolved to `\tmp` on the current drive. The `mkdir` runs at import time, so it raised during collection rather than in a test.

`tempfile.gettempdir()` already honours `TMPDIR` on POSIX and `TEMP`/`TMP` on Windows, so the behaviour is unchanged everywhere it previously worked.

**Worth flagging for anyone trying to reproduce this:** it does not fail on a machine that happens to have a `C:\tmp` directory. Mine did, and the first run collected 10 tests cleanly. I only saw the real failure after renaming that directory. If someone cannot reproduce the issue, that is the first thing to check.

## 2. `os.geteuid()` in `skipif` conditions

The issue names one occurrence. There are two, at lines 145 and 390. A `skipif` condition is evaluated when the class body executes, so the module cannot be imported on Windows regardless of what the condition would have returned.

The `@pytest.mark.skipif(os.name == "nt", ...)` decorator directly above each one does not help, for the same reason: both decorators' conditions are evaluated before either is applied.

The same file already uses the deferred form at line 293, with a comment explaining exactly this. These two were missed.

## Not included

`ruff check` passes on both files. `ruff format --check` reports these files as unformatted, but it did so before this change too, and reformatting them would bury a 19-line fix under several hundred lines of unrelated churn. Left alone deliberately.

The issue also suggests a Windows CI job or a `pytest --collect-only` smoke check. That seems right, and collection-time breakage is invisible to the per-file runner CI uses, but it is a separate change and I did not want to bundle it.

### DIFF
--- a/tests/hermes_cli/test_doctor_journal_modes.py
+++ b/tests/hermes_cli/test_doctor_journal_modes.py
@@ -142,7 +142,10 @@ class TestReadJournalMode:
             holder.close()
 
     @pytest.mark.skipif(os.name == "nt", reason="chmod is a no-op on Windows")
-    @pytest.mark.skipif(os.geteuid() == 0, reason="root ignores file permissions")
+    @pytest.mark.skipif(
+        getattr(os, "geteuid", lambda: 1)() == 0,
+        reason="root ignores file permissions",
+    )
     def test_read_only_directory_is_still_readable(self, tmp_path):
         db = tmp_path / "state.db"
         _make_db(db, journal_mode="WAL")
@@ -387,7 +390,10 @@ class TestReportDatabaseJournalModes:
         assert "state.db: rollback journal mode" in out
 
     @pytest.mark.skipif(os.name == "nt", reason="chmod is a no-op on Windows")
-    @pytest.mark.skipif(os.geteuid() == 0, reason="root ignores file permissions")
+    @pytest.mark.skipif(
+        getattr(os, "geteuid", lambda: 1)() == 0,
+        reason="root ignores file permissions",
+    )
     def test_unreadable_database_does_not_crash(self, tmp_path, capsys):
         db = tmp_path / "state.db"
         _make_db(db)
--- a/tests/test_run_tests_parallel.py
+++ b/tests/test_run_tests_parallel.py
@@ -24,6 +24,7 @@ import json
 import os
 import subprocess
 import sys
+import tempfile
 import textwrap
 import time
 from pathlib import Path
@@ -32,10 +33,16 @@ import pytest
 
 
 # Both tests share the same handoff file: the leaker writes here, the
-# verifier reads here. We park it in $TMPDIR with a unique-per-run name
-# so concurrent invocations of the suite don't clobber each other.
-_HANDOFF_DIR = Path(os.environ.get("TMPDIR", "/tmp")) / "hermes-isolation-probe"
-_HANDOFF_DIR.mkdir(exist_ok=True)
+# verifier reads here. We park it in the system temp dir with a
+# unique-per-run name so concurrent invocations of the suite don't
+# clobber each other.
+#
+# tempfile.gettempdir() already honours TMPDIR on POSIX and TEMP/TMP on
+# Windows. The previous hardcoded /tmp fallback resolved to \tmp on the
+# current drive on Windows, which does not exist, so this mkdir raised
+# at import time and aborted collection for the entire suite.
+_HANDOFF_DIR = Path(tempfile.gettempdir()) / "hermes-isolation-probe"
+_HANDOFF_DIR.mkdir(parents=True, exist_ok=True)
 
 
 def _handoff_path_for(nonce: str) -> Path:


alt-glitch:

Duplicate of #84073: both make the same doctor-journal geteuid and isolation-probe temp-directory collection paths portable on Windows.

monerostar:

Host: Windows 11 (build 26200), Python 3.11.15, pytest 9.1.1. Live collection, not a paper review.

### Before (current `main` files in the same tree)

`tests/hermes_cli/test_doctor_journal_modes.py` collection dies on the bare `os.geteuid()` skipif:

```
E   AttributeError: module 'os' has no attribute 'geteuid'
Interrupted: 1 error during collection
no tests collected, 1 error
```

`tests/test_run_tests_parallel.py` collection dies when the handoff parent is missing (`mkdir` without `parents=True`). This box has `C:\tmp`, so the `/tmp` fallback would not fail here (same caveat you called out). Forcing `TMPDIR` to a missing parent reproduces the class:

```
E   FileNotFoundError: [WinError 3] The system cannot find the path specified:
    '...\Temp\hermes-cron-no-such-parent-97801\hermes-isolation-probe'
Interrupted: 1 error during collection
```

That is why `python -m pytest` over the suite is unusable on Windows even when `scripts/run_tests_parallel.py` (one pytest per file, and often a real `TEMP`) stays green.

### After this PR

| file | collected |
|---|---|
| `test_doctor_journal_modes.py` | 36 |
| `test_run_tests_parallel.py` | 10 |

- Doctor file: **31 passed, 3 skipped, 2 failed** in 2.87s. The skips are the chmod tests (`os.name == "nt"`). The two failures are assertion portability, not this diff:
  - `test_missing_file_keeps_the_os_error_text` still wants POSIX `"No such file or directory"`; Windows returns `[WinError 2] The system cannot find the file specified: '...gone.db'`
  - `test_lists_every_managed_database` still wants `kanban/boards/myboard/kanban.db` (forward slashes); output uses `os.path.join` so `kanban\boards\myboard\kanban.db`
  - Those two are exactly https://github.com/NousResearch/hermes-agent/pull/97536 — this PR does not need to absorb them, but they should land together or collection-fixed Windows still cannot go green on this file.
- `test_run_tests_parallel.py -k "not grandchild_leak"`: **9 passed** in 11.52s.

`tempfile.gettempdir()` on this host: `C:\Users\Admin\AppData\Local\Temp`. `hasattr(os, "geteuid")` is False.

### Notes (non-blocking)

- Sibling overlap: https://github.com/NousResearch/hermes-agent/pull/74083 is the same isolation-probe handoff change. This PR is the one that also unblocks `test_doctor_journal_modes.py`, which is the bigger collection abort.
- Style nit only: one skipif in the same file already uses `hasattr(os, "geteuid") and os.geteuid() == 0`; the two you touched use `getattr(os, "geteuid", lambda: 1)() == 0`. Both are correct.

Fixes the reported #83935 collection abort on native Windows. Happy to see it land; #97536 is the follow-on so the newly collectable doctor tests actually pass here.

iamiskender:
Closing in favour of #84073, which does the same two fixes and was opened eighteen days earlier.

One thing from my testing that might be worth adding there: the /tmp failure does not reproduce on a machine that happens to have a C:\tmp directory. Mine did, and my first run collected 10 tests cleanly. I only saw FileNotFoundError after renaming that directory. Worth knowing if anyone cannot reproduce it.

PR #97803, opened 2026-08-29
fix(proxy): identify Nous requests as Hermes

### LINKED ISSUE #97796: fix(proxy): identify Nous subscription requests as Hermes
## Summary

The official `hermes proxy --provider nous` forwards the downstream client's `User-Agent` to the Nous Portal inference endpoint. OpenAI Python clients therefore identify the credential-attaching Hermes proxy as `OpenAI/Python ...`. The Portal's Cloudflare policy can reject that fingerprint with HTTP 403 / error 1010 before the request reaches inference.

## Production reproduction

Observed against the Nous Portal on 2026-08-06:

1. Send a completion through `hermes proxy --provider nous` using an OpenAI Python client.
2. The proxy forwards `User-Agent: OpenAI/Python ...` upstream.
3. The Portal returns HTTP 403 with Cloudflare error 1010.
4. Override only the upstream User-Agent to `HermesAgent/<version>`.
5. The same live completion succeeds with HTTP 200.

This is the same Cloudflare policy class already fixed for the `/v1/models` probe in #12999, but the subscription proxy has a separate forwarding path.

## Root cause

`hermes_cli/proxy/server.py` strips hop-by-hop and client authorization headers, then forwards every other client header unchanged. `NousPortalAdapter` has no way to declare provider-required request headers, so the downstream SDK fingerprint is sent to the Portal verbatim.

## Expected behavior

The credential-attaching proxy should identify itself upstream as Hermes, regardless of the downstream SDK. Client headers should remain unchanged for adapters that have no provider-specific requirement.

## Proposed fix

- Add an empty `upstream_headers` mapping to `UpstreamAdapter`.
- Have `NousPortalAdapter` return `User-Agent: HermesAgent/<version>`.
- Apply adapter headers after filtering client headers and before adding Hermes-owned authorization.
- Add an integration-style proxy test proving provider headers override client SDK fingerprints.

## Acceptance

- An `OpenAI/Python ...` downstream User-Agent is replaced by `HermesAgent/<version>` for the Nous upstream.
- Other adapters preserve current behavior by default.
- The regression test fails under the previous forwarding behavior and passes with the fix.

apellis:
## Summary

- let upstream adapters declare provider-required request headers
- identify Nous subscription proxy requests as `HermesAgent/<version>`
- make provider headers override downstream SDK fingerprints while preserving default adapter behavior

Fixes #97796.

## Problem

`hermes proxy --provider nous` attaches Nous credentials but forwards the downstream client's `User-Agent`. OpenAI Python callers therefore reach the Portal as `OpenAI/Python ...`, which the Portal's Cloudflare policy has rejected with HTTP 403 / error 1010. Changing only that upstream header made the same live completion succeed with HTTP 200.

This is the proxy-path sibling of #12999, which fixed the same Cloudflare policy class for the `/v1/models` probe.

## Approach

`UpstreamAdapter.upstream_headers` defaults to an empty mapping, so every existing adapter keeps current behavior. `NousPortalAdapter` supplies the Hermes User-Agent. The proxy applies those trusted adapter requirements after filtering client headers and before setting proxy-owned authorization.

## Verification

- `scripts/run_tests.sh -j 1 tests/hermes_cli/test_proxy.py -q` — 5 passed
- regression sabotage: restoring the old forwarding code makes `test_server_adapter_headers_override_client_headers` fail with `OpenAI/Python test`; restoring the fix returns to 5 passed
- `python -m compileall -q hermes_cli/proxy`
- `git diff --check`

## Risk

Low and provider-scoped. Other adapters return the empty default mapping. Nous requests change only the upstream User-Agent; request body, endpoint, credentials, and response streaming are unchanged.

### DIFF
--- a/hermes_cli/proxy/adapters/base.py
+++ b/hermes_cli/proxy/adapters/base.py
@@ -15,7 +15,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import FrozenSet, Optional
+from typing import FrozenSet, Mapping, Optional
 
 
 @dataclass(frozen=True)
@@ -58,6 +58,11 @@ class UpstreamAdapter(ABC):
         ``http://127.0.0.1:<port>/v1/chat/completions``. Requests to paths
         not in this set get a 404 with a helpful error body.
         """
+
+    @property
+    def upstream_headers(self) -> Mapping[str, str]:
+        """Provider-required request headers that override client headers."""
+        return {}
 
     @abstractmethod
     def is_authenticated(self) -> bool:
--- a/hermes_cli/proxy/adapters/nous_portal.py
+++ b/hermes_cli/proxy/adapters/nous_portal.py
@@ -11,6 +11,7 @@ import logging
 import threading
 from typing import Any, Dict, FrozenSet, Optional
 
+from hermes_cli import __version__ as HERMES_VERSION
 from hermes_cli.auth import (
     AuthError,
     DEFAULT_NOUS_INFERENCE_URL,
@@ -61,6 +62,13 @@ class NousPortalAdapter(UpstreamAdapter):
     @property
     def allowed_paths(self) -> FrozenSet[str]:
         return _ALLOWED_PATHS
+
+    @property
+    def upstream_headers(self) -> Dict[str, str]:
+        # The Portal's Cloudflare policy rejects SDK fingerprints such as
+        # ``OpenAI/Python`` with error 1010. Identify the credential-attaching
+        # proxy itself rather than forwarding the downstream client's UA.
+        return {"User-Agent": f"HermesAgent/{HERMES_VERSION}"}
 
     def is_authenticated(self) -> bool:
         state = self._read_state()
--- a/hermes_cli/proxy/server.py
+++ b/hermes_cli/proxy/server.py
@@ -142,6 +142,10 @@ def create_app(adapter: UpstreamAdapter) -> "web.Application":
                 upstream_url = f"{upstream_url}?{request.query_string}"
 
             fwd_headers = _filter_request_headers(request.headers)
+            # Provider requirements take precedence over downstream SDK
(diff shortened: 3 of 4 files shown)


alt-glitch:

Duplicate of #97804: both fix the Nous proxy User-Agent forwarding path for #97796. #97804 also removes case-insensitive downstream header collisions.

zodd7:
I independently reproduced this issue and can confirm the fix resolves the Nous 403 / Cloudflare 1010 rejection in live OAuth-proxy testing.

Reproduction: with an unpatched `hermes proxy --provider nous`, a downstream OpenAI-compatible client sending `User-Agent: OpenAI/Python ...` reached the Nous upstream unchanged and was rejected (HTTP 403 / error 1010). After overriding the upstream `User-Agent` to `HermesAgent/<version>`, the same request returned HTTP 200 — matching the diagnosis in the issue.

This PR's approach (a provider-owned `upstream_headers` mapping applied after client-header filtering and before the proxy-owned `Authorization`, with `NousPortalAdapter` supplying `HermesAgent/<version>`) is the right contract and is correctly scoped to the Nous adapter.

If it's helpful, I have a separate isolated reproduction with additional regression cases that may be worth folding into the test suite:

- downstream `OpenAI/Python` User-Agent -> upstream receives `HermesAgent/<version>`
- arbitrary downstream User-Agent -> upstream still receives `HermesAgent/<version>`
- downstream `Authorization` -> never forwarded upstream
- non-Nous adapter -> existing User-Agent behavior unchanged
- `/v1/models` -> still works
- streaming (SSE) -> still works
- OAuth refresh / runtime JWT rotation + restart persistence -> unchanged

The cases use a fake upstream that records the received request headers (non-streaming and SSE), so they assert the exact upstream `User-Agent` and that the client `Authorization` is replaced. Happy to contribute them as a patch to `tests/hermes_cli/test_proxy.py` if a maintainer would like the extra coverage — no pressure, and I won't open a competing PR.

PR #97804, opened 2026-08-29
fix(proxy): identify Nous subscription requests as Hermes

### LINKED ISSUE #97796: fix(proxy): identify Nous subscription requests as Hermes

liuhao1024:
## What does this PR do?

`hermes proxy --provider nous` forwarded the downstream client's `User-Agent` verbatim to the Nous Portal inference endpoint, so OpenAI SDK clients identified the credential-attaching proxy as `OpenAI/Python ...` — and the Portal's Cloudflare policy rejected that fingerprint with HTTP 403 / error 1010 before the request reached inference. This is the same Cloudflare policy class already fixed for the `/v1/models` probe in #12999, but the subscription proxy has a separate forwarding path.

The fix lets an adapter declare provider-required headers and applies them on the proxy's forwarding path, so the proxy identifies itself upstream as Hermes regardless of the downstream SDK:

- `UpstreamAdapter` gains an `upstream_headers` property, empty by default — adapters without provider-specific requirements keep forwarding client headers unchanged (e.g. the xAI adapter).
- `NousPortalAdapter` declares `User-Agent: HermesAgent/<version>`, mirroring the existing probe header in `hermes_cli/models.py`.
- The server applies adapter headers after client-header filtering and before adding the Hermes-owned `Authorization`. Declared keys are matched case-insensitively: a client may send `user-agent` in any casing, and a plain `dict.update` would forward both case variants upstream as duplicate headers.

## Related Issue

Fixes #97796

## Type of Change

## Changes Made

- `hermes_cli/proxy/adapters/base.py`: added `UpstreamAdapter.upstream_headers` property (default empty) + docstring responsibility note.
- `hermes_cli/proxy/adapters/nous_portal.py`: `NousPortalAdapter.upstream_headers` returns `User-Agent: HermesAgent/<hermes_cli.__version__>`.
- `hermes_cli/proxy/server.py`: added `_apply_upstream_headers()` and wired it into the forwarding path between `_filter_request_headers()` and the bearer rewrite.
- `tests/hermes_cli/test_proxy.py`: `FakeAdapter` accepts `upstream_headers`; the fake upstream's echo captures all `User-Agent` values (any casing) to catch duplicates; three new tests (see How to Test).

## How to Test

1. `pytest tests/hermes_cli/test_proxy.py -q` — all 7 tests pass.
2. `test_server_applies_adapter_upstream_headers_over_client_ua` — integration test through the real aiohttp proxy + fake upstream: a lowercase `user-agent: OpenAI/Python 4.104.0` client header is replaced by the adapter's `HermesAgent/test` value, and exactly one User-Agent reaches the upstream. Observed result: passes with the fix; fails under the previous forwarding behavior (verified by temporarily unwiring `_apply_upstream_headers`).
3. `test_server_preserves_client_ua_without_adapter_overrides` — adapters with no `upstream_headers` still forward the client's `User-Agent` unchanged. Observed result: passes.
4. `test_nous_adapter_declares_hermes_user_agent` — `NousPortalAdapter().upstream_headers` is exactly `{"User-Agent": "HermesAgent/<version>"}`. Observed result: passes.
5. Regression sweep over all test files importing `hermes_cli.proxy` (`test_proxy.py`, `test_iron_proxy_cli.py`, `test_nous_inference_url_validation.py`, `test_cli_status_command.py`, `test_unknown_command.py`): 40 passed.

### Code

### DIFF
--- a/hermes_cli/proxy/adapters/base.py
+++ b/hermes_cli/proxy/adapters/base.py
@@ -7,6 +7,7 @@ local proxy can forward requests to. The adapter is responsible for:
   - refreshing/minting credentials when needed
   - reporting the resolved upstream base URL
   - declaring which request paths it accepts
+  - declaring provider-required request headers, if any
 
 The proxy server is otherwise provider-agnostic.
 """
@@ -15,7 +16,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import FrozenSet, Optional
+from typing import Dict, FrozenSet, Optional
 
 
 @dataclass(frozen=True)
@@ -58,6 +59,18 @@ class UpstreamAdapter(ABC):
         ``http://127.0.0.1:<port>/v1/chat/completions``. Requests to paths
         not in this set get a 404 with a helpful error body.
         """
+
+    @property
+    def upstream_headers(self) -> Dict[str, str]:
+        """Request headers this provider requires on every forwarded request.
+
+        Applied after client headers are filtered and before the proxy adds
+        its own ``Authorization``, so a provider requirement (e.g. a
+        provider-specific ``User-Agent``) overrides the downstream SDK
+        fingerprint. Empty by default — adapters without provider-specific
+        requirements keep forwarding client headers unchanged.
+        """
+        return {}
 
     @abstractmethod
     def is_authenticated(self) -> bool:
--- a/hermes_cli/proxy/adapters/nous_portal.py
+++ b/hermes_cli/proxy/adapters/nous_portal.py
@@ -62,6 +62,20 @@ class NousPortalAdapter(UpstreamAdapter):
     def allowed_paths(self) -> FrozenSet[str]:
         return _ALLOWED_PATHS
 
+    @property
+    def upstream_headers(self) -> Dict[str, str]:
+        # The Portal's Cloudflare policy rejects SDK fingerprints like
+        # ``OpenAI/Python`` with HTTP 403 (error 1010) before inference sees
+        # the request, so the credential-attaching proxy must identify itself
+        # as Hermes regardless of the downstream client. Same policy class
+        # already fixed for the /v1/models probe (#12999).
+        try:
+            from hermes_cli import __version__
+            _v = str(__version__)
+        except Exception:
+            _v = "0"
+        return {"User-Agent": f"HermesAgent/{_v}"}
+
     def is_authenticated(self) -> bool:
         state = self._read_state()
         if state is None:
--- a/hermes_cli/proxy/server.py
+++ b/hermes_cli/proxy/server.py
@@ -72,6 +72,22 @@ def _filter_request_headers(headers: "aiohttp.typedefs.LooseHeaders") -> dict:
     return out
 
 
+def _apply_upstream_headers(forwarded: dict, overrides: dict) -> dict:
+    """Apply adapter-declared provider headers over the filtered client set.
+
+    With no overrides the client's headers pass through unchanged. Declared
+    keys are matched case-insensitively — a client may send ``user-agent``
+    in any casing, and a plain ``dict.update`` would forward both case
+    variants upstream as duplicate headers.
+    """
+    if not overrides:
+        return forwarded
+    overridden = {key.lower() for key in overrides}
+    out = {k: v for k, v in forwarded.items() if k.lower() not in overridden}
+    out.update(overrides)
+    return out
+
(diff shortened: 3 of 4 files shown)


Enough1122:

Reviewed `hermes_cli/proxy/adapters/base.py`, `hermes_cli/proxy/adapters/nous_portal.py`, `hermes_cli/proxy/server.py` + tests for PR #97804. The design is clean: a new `upstream_headers` property on the adapter ABC declares provider-required request headers, and `_apply_upstream_headers` applies them over the filtered client set *before* the proxy's own `Authorization`, with case-insensitive key matching so a lowercase `user-agent` from the client collapses into the single adapter value instead of forwarding two duplicate headers upstream. The Nous adapter returns `HermesAgent/<version>` to satisfy the Portal's Cloudflare policy (the same class as the `/v1/models` fix in #12999). Tests cover override, no-override passthrough, and the adapter declaration.

Non-blocking observations:

1. **The override removes the downstream client's identity fingerprint unconditionally.** Every client behind this proxy (SDK, MCP server, scripts) is now identified upstream as `HermesAgent/<version>`, regardless of what it sets. That's the intent (avoid the Cloudflare 403), but it also means upstream analytics/auditing can no longer distinguish real client types, and any client that *deliberately* sets a UA (e.g. for platform gating) silently loses it. The docstring documents the override behavior; just flagging the trade-off is real and applies to all traffic through the Nous proxy.
2. **`from hermes_cli import __version__` runs on every request.** `upstream_headers` is a property evaluated per forwarded request, and the import is inside it. A module-level lazy cache (or memoized property) would avoid the import cost per request — minor, since `import` is cached after first load, but the attribute-lookup + try/except per request is unnecessary.
3. **Duplicate-case overrides within the adapter dict itself aren't collapsed.** `_apply_upstream_headers` collapses case variants from the *client* set (`out = {k: v for k, v in forwarded.items() if k.lower() not in overridden}`), but `out.update(overrides)` copies the adapter's keys verbatim — if an adapter ever declared `{"User-Agent": ..., "user-agent": ...}`, both would be forwarded as duplicates. No current adapter does this, so it's theoretical; a `lower()`-keyed dict comprehension on `overrides` would make it bulletproof.
4. **Minor:** `_v = "0"` fallback on import failure produces `HermesAgent/0` — fine as a degraded identity, but a static literal like `HermesAgent/unknown` would be clearer in logs.

Solid, minimal, and well-tested; nothing blocking.

liuhao1024:
Thanks for the review. The UA-uniformity trade-off is real and is the documented price of the Portal's Cloudflare policy; the import-per-request and case-dedup notes are fair micro-hardenings I can fold in if wanted.

PR #97806, opened 2026-08-29
test(cli): cover ANSI color gating and application in hermes_cli.colors

salch-cred:
Add behavioral regression coverage for `hermes_cli.colors` — the ANSI color gating + application module used by ~20 modules across the CLI, which had **zero** direct unit coverage.

## Why

`colors.py`'s `should_use_color()` and `color()` gate every colored surface in the CLI (`NO_COLOR` env, `TERM=dumb`, TTY detection, code application). It's imported by ~20 files (`auth.py`, `config.py`, `doctor.py`, `curses_ui.py`, etc.) but has no direct tests — a subtle change to the gating logic (e.g. breaking the `NO_COLOR` check) would ripple across the whole CLI undetected.

## What

`tests/hermes_cli/test_colors.py` — behavioral tests (not change-detectors):

- `should_use_color()` returns False when `NO_COLOR` is set, when `TERM=dumb`, and when stdout is not a TTY.
- `color()` returns the text unchanged when color is disabled (no spurious ANSI escapes in piped output).
- `color()` applies the codes + reset when color is enabled (single and multiple codes).
- The stable ANSI SGR constants (`RESET`, `RED`, `GREEN`, `YELLOW`) are pinned as a contract.

## Validation

All assertions verified to pass against the current `colors.py` on `main`.

### DIFF
--- a/contributors/emails/salch-cred@users.noreply.github.com
+++ b/contributors/emails/salch-cred@users.noreply.github.com
@@ -1,0 +1,1 @@
+salch-cred
--- a/tests/hermes_cli/test_colors.py
+++ b/tests/hermes_cli/test_colors.py
@@ -37,10 +37,13 @@ def test_should_use_color_non_tty_disables(monkeypatch):
 
 
 def test_should_use_color_no_color_takes_precedence_over_tty(monkeypatch):
-    # Even with NO_COLOR set, non-TTY still yields False — but the point is
-    # the env gate applies regardless of the tty state.
+    # A real TTY would otherwise enable color, but NO_COLOR must override it.
+    class _Tty(io.StringIO):
+        def isatty(self):
+            return True
+
     monkeypatch.setenv("NO_COLOR", "1")
-    monkeypatch.setattr("hermes_cli.colors.sys.stdout", io.StringIO())
+    monkeypatch.setattr("hermes_cli.colors.sys.stdout", _Tty())
     assert should_use_color() is False
 
 
--- a/tests/hermes_cli/test_colors.py
+++ b/tests/hermes_cli/test_colors.py
@@ -1,0 +1,78 @@
+"""Behavioral tests for hermes_cli.colors — ANSI color gating + application.
+
+`colors.py` is imported by ~20 modules but had no direct unit coverage. These
+pin the *behavior* (when color is on/off and how it's applied), not values, so
+they don't break on cosmetic changes.
+"""
+
+import io
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed `tests/hermes_cli/test_colors.py` (new) for PR #97806. The tests correctly pin the *behavior* of `hermes_cli.colors` — `NO_COLOR`, `TERM=dumb`, non-TTY disabling, and code application/concatenation — rather than the internals of every call site. Using a `StringIO` for non-TTY and a `_Tty(StringIO)` subclass for TTY is a clean, hermetic way to exercise the gating without a real terminal. Adding the first direct coverage for a module imported by ~20 files is genuinely valuable.

Non-blocking observations:

1. **`test_should_use_color_no_color_takes_precedence_over_tty` doesn't test what its name claims.** It sets `NO_COLOR` + `StringIO` (non-TTY) and asserts `False` — but non-TTY already yields `False`, so this test would pass even if `NO_COLOR` were ignored entirely. The actual precedence claim ("NO_COLOR beats a TTY") needs a `_Tty` + `NO_COLOR` case asserting `False` (mirroring how `test_color_applies_codes_when_color_enabled` fakes a TTY). As written, the name overstates the coverage.
2. **`test_colors_constants_are_ansi_escapes` asserts literal escape values.** `Colors.RED == "\x1b[31m"` pins a specific value. The ANSI SGR sequences are a stable standard, so this won't break on routine data changes (not a classic change-detector test), but it is a snapshot of the constant values rather than behavior — if the module ever mapped colors to named aliases or offered a `Colors`-by-name lookup, a behavior-level assertion would be stronger. Acceptable as a documented contract.
3. **Minor:** all tests rely on `monkeypatch.setattr("hermes_cli.colors.sys.stdout", ...)`; if `should_use_color()` is ever refactored to read stdout through a helper or to cache its result, several tests would silently no-op. A tiny shared fixture (like `_Tty`) would reduce the repetition — cosmetic.

Valuable coverage addition; just the one misnamed test (#1) is worth tightening.

salch-cred:
Thanks — caught a real flaw. The precedence test was misnamed: it set NO_COLOR + non-TTY and asserted False, which passes regardless of NO_COLOR. Fixed it to fake a TTY with NO_COLOR set and assert False, so it actually pins 'NO_COLOR beats a TTY' (pushed on this branch). Also took the cosmetic note — the _Tty/StringIO repetition is now isolated where it matters; kept it inline to avoid over-abstracting a small module. The constant-pinning stays as a documented ANSI contract.

PR #97807, opened 2026-08-29
fix(delegation): size child context files by model window

ahaone:
## Summary
- propagate the delegated child model context window into project context-file loading
- prefer the constructed child compressor window, with best-effort delegated-model metadata fallback
- keep workspace context loading one-shot and preserve the existing `None`/exception fallback

## Test plan
- `scripts/run_tests.sh tests/tools/test_delegate.py -q` (74 passed)

## TDD receipt
- RED: 4 new focused regressions failed before the implementation
- GREEN: the focused regressions and the complete delegate test file pass

### DIFF
--- a/tests/tools/test_delegate.py
+++ b/tests/tools/test_delegate.py
@@ -147,6 +147,97 @@ class TestChildSystemPrompt(unittest.TestCase):
         self.assertIn("YOUR TASK", prompt)
         self.assertNotIn("CONTEXT", prompt)
 
+    @patch("agent.prompt_builder.build_context_files_prompt")
+    def test_forwards_explicit_context_length(self, mock_context_prompt):
+        mock_context_prompt.return_value = "project conventions"
+
+        prompt = _build_child_system_prompt(
+            "Fix the tests",
+            workspace_path="/tmp/workspace",
+            context_length=131072,
+        )
+
+        mock_context_prompt.assert_called_once_with(
+            cwd="/tmp/workspace", skip_soul=True, context_length=131072
+        )
+        self.assertIn("project conventions", prompt)
+
+
+class TestChildPromptContextLength(unittest.TestCase):
+    def _build_with_context_length(self, compressor_length, metadata_result=None, metadata_error=None):
+        parent = _make_mock_parent()
+        child = MagicMock()
+        child.context_compressor.context_length = compressor_length
+
+        def construct_child(**kwargs):
+            self.assertNotIn("project-context-", kwargs["ephemeral_system_prompt"])
+            return child
+
+        metadata = MagicMock()
+        if metadata_error is not None:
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed `tools/delegate_tool.py` + tests for PR #97807. The change sizes delegated child context-file loading by the child's actual model window: `_build_child_system_prompt` gains a `context_length` parameter forwarded to `build_context_files_prompt`, and `_build_child_agent` resolves it *after* constructing the child — preferring the child compressor's `context_length` (strict `type(...) is int and > 0`, which correctly excludes `bool` and `0`), falling back to `get_model_context_length` metadata, then `None` (original behavior) on failure. The two-stage prompt build (bootstrap without workspace context, then `setattr(child, "ephemeral_system_prompt", ...)` with the resolved length) is a sensible chicken-and-egg solution since the child's real window is only known post-construction. Tests cover explicit forwarding, compressor-priority (metadata not called), invalid-compressor fallback, and metadata-failure → `None`.

Non-blocking observations:

1. **The first-stage prompt is constructed and then immediately replaced.** `_build_child_system_prompt(goal, context, role=..., ...)` is built *without* `workspace_path`/`context_length` and passed into `AIAgent(...)` construction, then `ephemeral_system_prompt` is overwritten with the second-stage prompt. If `AIAgent.__init__` uses the passed `ephemeral_system_prompt` for anything cached (token estimation, prompt-cache key derivation, logging) rather than only reading it later, the overwrite could leave that derived state inconsistent. The tests assert the *kwargs* prompt lacks `project-context-` but don't verify `AIAgent` never uses the first prompt after construction. Worth a comment in `_build_child_agent` (or a check) confirming the first prompt is purely a bootstrap that is never consumed post-init.
2. **`setattr(child, "ephemeral_system_prompt", ...)` bypasses any setter/hook.** Direct attribute assignment on the `AIAgent` instance. If `ephemeral_system_prompt` is read at session/agent start via a property or requires invalidation of derived state, a plain `setattr` could silently diverge. Given the existing code already treats it as a plain attribute, this is consistent — just noting the overwrite is intentionally low-level.
3. **Metadata fallback can be a network/slow call per delegated child.** `get_model_context_length(model, base_url=..., api_key=..., provider=...)` is invoked when the child compressor lacks a valid length. If that function ever performs a remote catalog lookup, it now runs once per child spawn instead of being resolved from the compressor. From the tests it behaves as a deterministic lookup, so likely fine — worth confirming it's cache-backed to avoid per-delegation latency on large batch spawns.
4. **Minor:** `type(compressor_length) is int` also rejects `numpy.int64` or other int-like values a compressor might return; consistent with the strict-typing choice, but a `numbers.Integral` check would be more forgiving if a compressor ever yields a non-`int` numeric type.

Clean, well-tested sizing fix; nothing blocking.

PR #97808, opened 2026-08-29
fix(desktop): surface persistent group holds

### LINKED ISSUE #97740: [Bug] Bot Mode group chat: sticky Stop holds are invisible in the room and `@all` without "resume" can't release them — rooms go permanently silent
## Bug Description

After a user clicks the room **Stop** button (shipped in #96241), every member gets a **persistent hold** (#93129). This is correct as a contract, but three properties combine into a trap where the room goes **permanently silent** with no visible cause:

1. **Holds are invisible in the room.** The only trace is a transient `held` event in the collapsible Activity panel (`group-activity.ts`: `"is held (stopped by you) — @mention it or say resume to release"`). The room transcript itself shows nothing: no banner, no system line, no member marker. Held turns are skipped **silently** — the user sees "message sent, nobody replies, not even thinking".

2. **Activity events are wiped by the next send.** Events are tagged with the room epoch and the view shows only the current run (`currentGroupActivity` filters `event.epoch === room.epoch`). Every new user send bumps the epoch (`sendToGroupChat` → `room.epoch += 1`), so the `held` explanation **disappears exactly when the user sends the next message that gets silently skipped**. The one diagnostic that explains the silence is destroyed by the act of probing the silence.

3. **`@all <anything except resume>` cannot release holds.** Hold release only triggers on `@all resume` or a direct non-stop mention of the held member (`classifyGroupHoldDirective` in `group-rounds.ts`: `everyone=true` with no `resume` word → `releaseAll: false`, `mentioned: ∅` → nothing released). A user trying to wake the room with `@all 你们依次给我说个笑话` gets **another silent round**: responders = everyone, every one of them held, all skipped. The room looks dead.

Additionally, holds are **persisted** (localStorage `group-chats`, by design — `updateGroupChat` stores `holds` so a restart doesn't free a bot the user explicitly stopped). That's the right durability choice, but it means neither app restart nor desktop update clears the state — the trap survives for days. We reproduced a room held silent across a full app update + restart (2026-08-28 → 08-29, commit `9a1eef7a2`).

## Steps to Reproduce

1. Bot Mode group room, ≥2 members (we used 3: Hermes/Reader/Docker, local profiles on Windows 10).
2. While a member turn is visibly running ("X is thinking…"), click **Stop**. (Easy to do accidentally: with poll-driven turns — #92760 — a 37–183 s joke round looks stuck.)
3. Send any message: `hello` → silence. Send `@all <task>` → silence. Send `@all` + explicit instructions → still silence. No errors anywhere (`errors.log` clean in the send window).
4. Open the collapsed Activity panel: a single `held` row explains it — but (2) means if you sent anything after the Stop, that row is already gone.
5. Send `@all resume` → next `@all` message works instantly: all members take turns, @mention handoff relay works end-to-end.

Live-verified end-to-end on 2026-08-29: `@all resume` → `@all 你们依次给我说个笑话` → Hermes (+183 s) → Reader (+21 s, handoff via @mention) → Docker (+37 s) → next round settles with `(pass)`. The engine itself is healthy; the failure mode is purely the invisible sticky hold.

## Expected Behavior

- The room should **surface held members persistently**: e.g. a muted "⏸ Reader, Docker are held (stopped) — say `@all resume`" line pinned under the composer, or a system entry in the transcript. It should not live only in an epoch-scoped activity feed.
- The first silent skip after a hold should be visible **in the room**, not only in the feed (the code already records `held` once per hold — render it where the user is looking).
- Consider an escape hatch: `@all` addressed at the whole room could be treated as intent-to-engage (release holds), or at least the room could reply with a visible "members are held" notice instead of settling silently.
- (Related friction, same area: `stop`/`halt`/`pause` next to a mention holds members — conservative by design per #93129, but non-English rooms get held by English keyword false-positives and can't see why.)

## Actual Behavior

Room settles silently. The user's only recovery paths are knowing the incantation `@all resume` or reading a transient activity event that the next send erases. We watched a user (a hermes-agent contributor) lose ~24 h of room availability to this across an update + restart.

## Environment

- OS: Windows 10 (desktop app, local profiles)
- hermes-agent commit `9a1eef7a2` (post hermes-bots modularization)
- Related/overlapping: #94726 §5 (group chat engine), #93129 (holds), #96241 (Stop button), #92760 (poll-driven slowness that invites the Stop click)

## Suggested Fix Direction

Minimal, class-level: persist a `heldBy` summary on the room record and render a persistent room-level indicator (banner/system line) listing held members with the release incantation; keep the epoch-scoped activity feed for history. Optionally: `@all` (bare or with a task) releases holds, since addressing the whole room is unambiguous intent to engage everyone.

sambai-dev:
## What does this PR do?

Sticky Group Chat Stop holds intentionally survive later turns and app reloads, but the Activity feed is run/epoch-scoped. That lifetime mismatch can reopen a room with every bot silently paused and no durable explanation.

This PR projects the canonical `room.holds` map into a persistent, accessible status row in the real Group Chat workspace. It names partially held bots, summarizes an all-held room, and explains the existing release gestures (`@member` or `@all resume`) without changing hold authority or mutation semantics.

## Related Issue

Fixes #97740

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- Add `GroupHoldStatus`, a flat room-level status derived only from persisted holds.
- Distinguish all-members holds from partial holds and retain a readable fallback for source-qualified keys while roster data is unavailable.
- Add release guidance plus English, Japanese, Simplified Chinese, and Traditional Chinese copy.
- Wire the status between the Group Chat header and run-scoped Activity panel.
- Add component, canonical-resume, source-scoped hydration-shape, and real-workspace regression coverage.

## How to Test

1. Create a Group Chat with at least two bots.
2. Send `@all stop`; confirm the room shows that all bots are paused and explains `@all resume`.
3. Reopen/reload the room; confirm the status remains even though old Activity entries do not.
4. Send `@all resume` or mention a paused bot; confirm the status clears through the existing hold directive.

Automated verification on Windows 11:

- `npm run test:ui -- src/plugins/hermes-bots/group-hold-status.test.tsx src/plugins/hermes-bots/group-rounds.test.ts src/plugins/hermes-bots/group-activity.test.ts src/plugins/hermes-bots/group-chat-view.test.ts` — **68 passed**
- `npm run test:ui -- src/plugins/hermes-bots` — **562 passed**
- `npm run check:lint` — **passed** (0 errors; existing repository warnings are outside changed files)
- `npm run build` — **passed**, including renderer, Electron main/preload, native staging, and dist assertion
- Regression sabotage: removing the workspace projection makes the selected test fail with no accessible `status`; restoring it passes

Full renderer disclosure (`npm run test:ui`, default parallelism):

- Final candidate: **6,719 passed, 9 failed**
- Clean `origin/main@aff5125f`: **6,714 passed, 8 failed**
- Every failure is in the same three unrelated files; the candidate's one additional failure is a timeout in the already baseline-failing `MessagingView` file, not a new failing file
- Those three files pass **12/12** together in isolation on the final candidate

Desktop platform disclosure (`npm run test:desktop:platforms` on Windows):

- Candidate: **1,909 passed, 46 failed**
- Clean `origin/main@aff5125f`: **1,908 passed, 47 failed**
- Every candidate failure is also present on baseline, in untouched POSIX/macOS filesystem/path-emulation or temporary-repository tests

Neither full-suite result is reported as green; both comparisons show no new failure from this renderer-only diff.

## Risk and Scope

Presentation-only. This PR does not change hold classification, persistence, release, routing, orchestration, or Activity retention. Removing `GroupHoldStatus` reverts the UI behavior.

### Code

- [ ] I've run `pytest tests/ -q` and all tests pass — N/A: Desktop TypeScript-only change; relevant gates are listed above
- [x] I've tested on my platform: Windows 11 x64

- [x] Documentation update — N/A; no public API or configuration changed
- [x] `cli-config.yaml.example` update — N/A; no config keys changed
- [x] `CONTRIBUTING.md` / `AGENTS.md` update — N/A; no architecture or workflow changed
- [x] Cross-platform impact considered; renderer-only React/TypeScript with no platform branch
- [x] Tool descriptions/schemas update — N/A; no tool behavior changed

## Screenshots / Logs

No screenshot is attached: the isolated Windows Desktop E2E sandbox exceeded its existing 90-second `beforeAll` boot ceiling before the visual test body ran, so no synthetic screenshot was substituted. The production build and real-workspace DOM regression both pass as documented above.

### DIFF
--- a/apps/desktop/src/plugins/hermes-bots/group-chat-view.tsx
+++ b/apps/desktop/src/plugins/hermes-bots/group-chat-view.tsx
@@ -72,6 +72,7 @@ import {
 import type { GroupChatRoom } from './group-chat'
 import { GroupClarifyCard, GroupImageControls, GroupMentionInput } from './group-chat-parts'
 import type { GroupRoomPrompt } from './group-chat-parts'
+import { GroupHoldStatus } from './group-hold-status'
 import {
   botGroups,
   groupChatMemberBots,
@@ -1233,6 +1234,11 @@ export function GroupChatWorkspace({ group, members, onBack, visible = true }: G
         </div>
       ) : null}
       {header}
+      <GroupHoldStatus
+        holds={room.holds}
+        memberLabel={member => displayName(member, botRosterMeta(member, allMeta))}
+        members={members}
+      />
       {activityPanel}
       <div className="min-h-0 flex-1 overflow-y-auto overscroll-contain">
         <div className="grid gap-1.5 px-2.5 pb-2">
--- a/apps/desktop/src/plugins/hermes-bots/group-hold-status.test.tsx
+++ b/apps/desktop/src/plugins/hermes-bots/group-hold-status.test.tsx
@@ -1,0 +1,166 @@
+import { cleanup, render, screen } from '@testing-library/react'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { translateBots } from './i18n-test-helper'
+import type { GroupMember } from './types'
+
+const { host } = vi.hoisted(() => ({ host: {} as Record<string, unknown> }))
+
+vi.mock('@hermes/plugin-sdk', async () => {
+  const { pluginSdkMock } = await import('./group-test-utils')
+  const base = await pluginSdkMock(host)
+
+  return {
+    ...base,
+    Button: (props: React.ComponentProps<'button'>) => <button type="button" {...props} />,
+    cn: (...values: unknown[]) => values.filter(Boolean).join(' '),
+    Codicon: ({ name }: { name: string }) => <span aria-hidden data-icon={name} />,
+    ConfirmDialog: () => null,
+    CopyButton: () => null,
+    Dialog: () => null,
+    DialogContent: ({ children }: { children?: React.ReactNode }) => <>{children}</>,
+    DialogDescription: ({ children }: { children?: React.ReactNode }) => <>{children}</>,
+    DialogFooter: ({ children }: { children?: React.ReactNode }) => <>{children}</>,
+    DialogHeader: ({ children }: { children?: React.ReactNode }) => <>{children}</>,
+    DialogTitle: ({ children }: { children?: React.ReactNode }) => <>{children}</>,
+    Input: (props: React.ComponentProps<'input'>) => <input {...props} />,
+    relativeTime: () => 'now',
+    RowButton: (props: React.ComponentProps<'button'>) => <button type="button" {...props} />,
+    Tip: ({ children }: { children?: React.ReactNode }) => <>{children}</>,
+    useI18n: () => ({ t: { common: { cancel: 'Cancel', save: 'Save' } } }),
+    usePluginI18n: () => translateBots
+  }
+})
+
+vi.mock('./group-chat-parts', () => ({
+  GroupClarifyCard: () => null,
+  GroupImageControls: () => null,
+  GroupMentionInput: (props: { 'aria-label'?: string }) => <textarea aria-label={props['aria-label']} />
+}))
+
+const MEMBERS: GroupMember[] = [
+  { name: 'research', title: 'Research' },
+  { name: 'builder', title: 'Builder' }
+]
+
+beforeEach(() => {
+  vi.resetModules()
+  Object.defineProperty(Element.prototype, 'scrollIntoView', {
+    configurable: true,
+    value: vi.fn()
+  })
+})
+
+afterEach(() => {
+  cleanup()
+  vi.clearAllMocks()
+})
+
+describe('durable group holds', () => {
+  it('shows an accessible all-members status without relying on the activity feed', async () => {
+    const { GroupHoldStatus } = await import('./group-hold-status')
+
+    render(
+      <GroupHoldStatus
+        holds={{ builder: { at: 2 }, research: { at: 1 } }}
+        memberLabel={member => member.title || member.name}
+        members={MEMBERS}
+      />
+    )
+
+    const status = screen.getByRole('status')
+
+    expect(status.textContent).toContain('All 2 bots are paused')
+    expect(status.textContent).toContain('Mention a paused bot or send @all resume to release them.')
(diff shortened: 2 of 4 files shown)


Enough1122:

Reviewed `apps/desktop/src/plugins/hermes-bots/group-hold-status.tsx` (+ view integration + i18n + tests) for PR #97808. The approach is sound: projecting the durable `room.holds` map into a `role="status"` row keeps the explanation alive across run/epoch reloads, and the tests pin the all-held, partial, source-qualified, and hydrated-workspace cases. i18n coverage across all four shipped locales is complete.

Non-blocking observations:

1. **Member↔hold matching relies entirely on `groupMemberKey` producing the same key shape the hold map uses.** The component splits `key.split('::')` only for *unmatched* keys (the `unavailableHeldLabels` path), so if `groupMemberKey(member)` ever emits an unqualified name (`builder`) while a persisted hold is source-qualified (`mac-mini::builder`) — or vice versa — the same bot can appear twice (once as a held member, once in the "unavailable" list). The tests cover all-unqualified and all-unqualified-absent but not the mixed case where a member row exists under a different key form than the hold. Worth one test asserting the "held member present + source-qualified hold for the same bot" display, or a comment documenting why `groupMemberKey` and hold keys are guaranteed identical.

2. **Minor:** the `allHeld` branch depends on `props.members.length > 0` and `unavailableHeldLabels.length === 0`. A fully-held room whose holds are all source-qualified while `members` is populated would render "Paused: …" rather than "All N bots are paused". Cosmetic only.

3. **Minor:** the component re-creates `Set` objects on every render and is invoked unconditionally in the workspace. For typical group sizes this is negligible; no action needed unless groups grow large.

The change is well-scoped, tested, and consistent with the existing group-chat parts. Nothing blocking.

teknium1:
Merged via PR #99094 — your commit was cherry-picked onto current main with your authorship preserved in git log (merge commit a9c783f219). All 6 tests pass including the real-workspace hydration case. Note: this covers the visibility half of #97740; the `@all`-without-"resume" release gap stays open there. Thanks!

PR #97813, opened 2026-08-29
fix(desktop): summarize multi-file patch cards

apellis:
## Summary

- show the first touched filename on pending V4A patch cards
- summarize additional files for multi-file pending and completed patches
- localize the additional-file count in all shipped Desktop locales
- ignore marker-looking prose outside a real `*** Begin Patch` boundary

## Problem

Desktop patch cards can render raw patch arguments or a generic label while a V4A patch is pending. Multi-file results also collapse to a single path, hiding the actual scope of the edit.

## Approach

The fallback model now collects unique paths from, in order:

1. an explicit single-file argument;
2. structured `files_modified` / `files_created` / `files_deleted` results;
3. a bounded V4A patch body;
4. the existing inline-diff fallback.

The visible title stays compact (`demo.ts +2 files`) while the subtitle retains the first full path. Existing single-file behavior is unchanged.

## Verification

- `npm --prefix apps/desktop run test:ui -- src/components/assistant-ui/tool/fallback-model.test.ts` — 42 passed
- full Desktop TypeScript typecheck passed
- scoped ESLint passed with 0 errors and 0 warnings
- regression sabotage: restoring substring boundary matching and English fallback makes the exact whole-line and Arabic tests fail; restoring the fix returns to 42 passed
- `git diff --check`

## Risk

UI presentation only. Patch execution and parsing are untouched. Missing or malformed boundaries fall back to the existing generic card instead of guessing filenames.

### DIFF
--- a/apps/desktop/src/components/assistant-ui/tool/fallback-model.test.ts
+++ b/apps/desktop/src/components/assistant-ui/tool/fallback-model.test.ts
@@ -216,6 +216,143 @@ describe('buildToolView file edit diffs', () => {
     expect(view.subtitle).toBe('src/demo.ts')
     expect(view.detail).toBe('')
   })
+
+  it('extracts the filename from a pending single-file V4A patch', () => {
+    const patch = [
+      '\x2a\x2a\x2a Begin Patch',
+      '\x2a\x2a\x2a Update File: src/components/demo.ts',
+      '@@',
+      '-old',
+      '+new',
+      '\x2a\x2a\x2a End Patch'
+    ].join('\n')
+
+    const view = buildToolView(
+      part({ args: { mode: 'patch', patch }, result: undefined, toolName: 'patch' }),
+      patchDiff
+    )
+
+    expect(view.title).toBe('demo.ts')
+    expect(view.subtitle).toBe('src/components/demo.ts')
+  })
+
+  it('summarizes all files represented by a multi-file V4A patch', () => {
+    const patch = [
+      '\x2a\x2a\x2a Begin Patch',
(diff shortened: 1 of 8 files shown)


Enough1122:

Reviewed `apps/desktop/src/components/assistant-ui/tool/fallback-model/index.ts` (+ i18n + tests) for PR #97813. The refactor from `fileEditPath` (single string) to `fileEditPaths` (array) with the V4A boundary parser is clean, and the tests thoroughly pin the multi-file count, localized strings, and the whole-line `*** Begin/End Patch` boundary rules (including ignoring marker-looking prose outside a real boundary).

Non-blocking observations:

1. **`v4aFilePaths` parses only the FIRST `*** Begin Patch` … `*** End Patch` block.** If a real V4A payload ever contains multiple independent Begin/End blocks, files in later blocks are silently dropped from the summary. The current tool contract appears to emit one block per patch, so this is likely fine — a comment noting the single-block assumption would make it explicit.

2. **A `Move File: A -> B` pushes both `A` and `B` into the path list** before dedup, so a rename is reported as `+1 file` (2 unique paths → `additional = 1`). Semantically a rename is one logical file; if the count is meant to be "files touched", that's correct, but it's worth confirming the intent (a rename showing "+1 file" is defensible either way).

3. **Minor:** the first file (not basename) drives both title and subtitle, and `fileEditTitle` uses `paths[0]` while subtitle presumably uses `fileEditPath` (same first element) — consistent, good. The `additionalFiles` i18n strings were added to all five locales including Arabic plural forms, which is thorough.

The change is well-scoped and the parser has good negative coverage. Nothing blocking.

PR #97814, opened 2026-08-29
fix(projects): report whether workspace moved

apellis:
## Summary

- report whether `desktop_project create` / `switch` actually moved the live chat workspace
- return `moved: false` plus a truthful warning for empty projects or unavailable sessions
- make the gateway workspace callback return its real outcome
- preserve legacy callbacks that return `None` after a successful move

## Problem

The project tools currently return success after updating the durable project DB even when the live chat cannot move—for example, an empty project has no folder, or the task no longer names a live gateway session. The caller can therefore claim that the chat moved when only the active-project pointer changed.

## Approach

The durable project operation remains successful. The response now distinguishes that from the live workspace side effect with a `moved` boolean and warning. The consolidated `desktop_project` description also states that empty projects cannot receive a chat.

## Verification

- `scripts/run_tests.sh -j 2 tests/tools/test_project_tools.py tests/tui_gateway/test_projects_rpc.py -q` — 36 passed
- regression sabotage: restoring the old project/gateway source makes the move-result tests fail; removing the finalized-session guard separately mutates the dead record and fails its regression test
- `python -m compileall -q tools/project_tools.py tui_gateway/server.py`
- `git diff --check`

## Risk

Backward-compatible response expansion: existing success/id/name/path fields remain. The only callback contract change is a boolean result from the built-in gateway callback; legacy `None` callbacks remain treated as successful.

### DIFF
--- a/tests/tools/test_project_tools.py
+++ b/tests/tools/test_project_tools.py
@@ -1,0 +1,47 @@
+import json
+
+from hermes_cli import projects_db
+from tools import project_tools
+
+
+def test_project_create_without_path_reports_that_chat_was_not_moved(monkeypatch, tmp_path):
+    monkeypatch.setattr(projects_db, "projects_db_path", lambda: tmp_path / "projects.db")
+    callback_calls = []
+    project_tools.set_project_workspace_callback(lambda *args: callback_calls.append(args) or True)
+
+    result = json.loads(project_tools.project_create("Empty", task_id="session-1"))
+
+    assert result["success"] is True
+    assert result["moved"] is False
+    assert "not moved" in result["warning"].lower()
+    assert callback_calls == []
+
+
+def test_project_switch_to_empty_project_reports_that_chat_was_not_moved(monkeypatch, tmp_path):
+    monkeypatch.setattr(projects_db, "projects_db_path", lambda: tmp_path / "projects.db")
+    callback_calls = []
+    project_tools.set_project_workspace_callback(lambda *args: callback_calls.append(args) or True)
+    with projects_db.connect_closing() as conn:
+        projects_db.create_project(conn, name="Empty")
+
+    result = json.loads(project_tools.project_switch("empty", task_id="session-1"))
+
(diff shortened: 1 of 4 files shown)


Enough1122:

Reviewed `tools/project_tools.py` + `tui_gateway/server.py` + tests for PR #97814. Making the workspace-move outcome explicit (`moved: bool` + truthful `warning`) is the right fix for the silent "DB updated, chat not moved" case, and the legacy-`None`-callback handling (`result is not False`) preserves back-compat cleanly. The `_finalized` skip in `_apply_project_workspace` (moving only a live session, not a finalized one) is a sensible refinement and is covered by tests.

Non-blocking observations:

1. **`_apply_workspace` treats any callback return other than `False` as "moved"** (`result is not False`). A callback that returns `0`/`''`/`None` is read as success — intended for legacy `None`, but it also silently swallows a buggy callback that returns a falsy non-`None`. Strictly, `is True` would be more precise, though the legacy-`None` contract documented in the diff justifies the current choice. Worth a one-line comment that only explicit `False` means "not moved".

2. **`_apply_workspace` returns `False` when `path` is falsy**, which is what produces the "no folder" warning — but it also returns `False` on any callback exception (swallowing `Exception`). That means a transient failure inside the gateway callback is reported to the user as "live chat workspace could not be moved" — accurate, but the original `pass`-and-succeed behavior (DB-only success) is now a hard failure signal. Fine for truthfulness; just confirming the intended trade-off (a callback hiccup now surfaces as `moved: false` + warning rather than silent DB-only success).

3. **Minor:** `_find_live_session_by_key` is referenced by `_apply_project_workspace`; it exists in `tui_gateway/server.py` (confirmed in the current tree), so the wiring is valid. The direct-key lookup now also checks `_finalized` before preferring a matching live session — covered by `test_project_workspace_callback_skips_finalized_session_for_live_continuation`.

The change is small, truthful, and well-tested. Nothing blocking.

PR #97811, opened 2026-08-29
fix(desktop): merge todo patches instead of replacing the Tasks list

Adolanium:
## What does this PR do?

`todo` with `merge: true` updates items by id on the Python store. The desktop Tasks panel treated `tool.start` `args.todos` as a full replace. A one-item patch became `Tasks 1/1`. A status-only patch (id and status, no content) parsed to an empty list, so the panel vanished until `tool.complete` shipped the full store. The in-progress spinner also hid unless the group was collapsed, which made `Tasks 0/5` look stuck.

`tool.start` now merges by id when `args.merge` is true. Status-only items still apply. The full result still replaces. The running spinner sits next to the header while the list is open.

## Related Issue

Fixes #97810

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `apps/desktop/src/lib/todos.ts`: `nextTodosFromToolEvent` prefers the full result, then merges `args` by id when `merge: true`. Status-only patches keep existing content.
- `apps/desktop/src/app/session/hooks/use-message-stream/index.ts`: live `todo` events go through that helper.
- `apps/desktop/src/components/chat/status-section.tsx`: show the running indicator next to the header even when the group is expanded.
- `apps/desktop/src/lib/todos.test.ts` and `apps/desktop/src/app/chat/composer/status-stack/collapsed-indicator.test.tsx`: merge, status-only patch, and expanded-header spinner.

## How to Test

1. From `apps/desktop`: `npx vitest run src/lib/todos.test.ts src/app/chat/composer/status-stack/collapsed-indicator.test.tsx`
2. `nextTodosFromToolEvent` with a five-item current list and `args: { merge: true, todos: [{ id: "c", status: "completed" }] }` must return five items with `c` completed.
3. In the desktop app, a live `todo` merge that marks one item in_progress must keep the other rows and show the header spinner. Completing items must move `Tasks 0/5` to `1/5`, then `2/5`, not jump to `5/5` at the end.

Ran on Windows 11: 16 passed in that filter. Live-checked in the desktop client on this branch.

### Code

[...]

Issue #19477, opened 2026-05-04
Todo tool state lost on gateway restart — no persistence mechanism


blakexu17:
## Bug Description

The `todo` tool (`tools/todo_tool.py`) stores all task data in an in-memory `TodoStore` instance (`self._items: list = []`). There is **no persistence to `state.db` or any file**. When the gateway process restarts or a new session starts, all todo items are silently lost.

This has caused **two data loss incidents** for a real user:
1. **2026-04-21 → 04-22**: Gateway restart lost 14 active todos. Detected 28 hours later, manually recovered from a markdown backup file.
2. **2026-04-23 → 2026-05-04**: Gateway restart again lost 16 todos. Detected by the user, manually recovered from `todos.json` (a backup file created after the first incident).

## Root Cause

In `tools/todo_tool.py`, `TodoStore.__init__` creates an empty list with no loading from any persistent store:

```python
class TodoStore:
    def __init__(self):
        self._items: List[Dict[str, str]] = []  # gone on restart
```

The `state.db` SQLite database has no todo table. The `AIAgent` creates a fresh `TodoStore` per session with no recovery logic.

## Steps to Reproduce

1. Add todos via the `todo` tool in a session
2. Restart the gateway (or wait for the next session)
3. Call `todo()` with no params → returns 0 items

## Expected Behavior

Todo state should survive gateway restarts. At minimum, auto-save to a JSON file on every write and auto-load on `__init__`.

## Proposed Fix

Add file-backed persistence to `TodoStore`:

1. **On every `write()` call**: serialize `self._items` to a JSON file (e.g. `~/.hermes/todos.json`, profile-aware)
2. **On `__init__()`**: if the JSON file exists, load items from it
3. **Fallback**: also write to `state.db` if a todo table exists

This is a minimal, backward-compatible change. The JSON persistence eliminates the data loss window entirely.

## Environment

- Hermes Agent: current main branch
- Profile: `~/.hermes-personal/`
- Platform: Telegram gateway
- OS: WSL (Ubuntu)

blakexu17:
## Additional Context: Existing Backup Files

The user profile already has two persistent backup files that work as a recovery path:

1. `~/.hermes-personal/todos.json` — JSON format, primary backup
2. `~/.hermes-personal/memories/bootstrap/sources/memory__personal_todos.md` — Markdown format, secondary fallback

A render script (`render_personal_todos.py`) already reads from these files with fallback logic:
```python
[...]

PR #97815, opened 2026-08-29
fix(todo): keep task state live across merges and reconnects

itsflownium:
## Summary

- persist full, revisioned todo snapshots in the existing session database
- emit a provider-neutral `todo.updated` event independently of optional tool-progress display settings
- restore authoritative todo state through session resume/activation and reject stale renderer updates
- merge `merge: true` live patches by task id, including status-only updates, without hiding the rest of the list
- keep the running indicator visible while the Tasks section is expanded

## Why

The Tasks panel currently depends on the optional `tool.complete` projection. A suppressed or missed event, renderer restart, or reminted runtime can therefore leave the panel stale even though the canonical tool result is correct.

This change keeps the transcript as an audit/fallback source while adding a compact session snapshot for live reconciliation. Compression continuations and branches inherit the nearest snapshot copy-on-write, so sibling branches remain isolated. Completion is never guessed from assistant text.

This is provider- and model-independent and does not change the todo tool schema presented to models beyond returning its monotonic revision.

This also incorporates the focused desktop fix from #97811 by @Adolanium, preserving its original commit and authorship.

## Verification

- backend regression suite: 30 passed
- combined desktop todo regression suite: 33 passed
- full desktop TypeScript typecheck
- production desktop build
- isolated Hermes desktop acceptance run: a 10-step process updated the Tasks panel live (captured at 4/10) and finished with 10/10 completed in the durable snapshot

Addresses #19477.

## Desktop acceptance

The live Tasks panel after four of ten deterministic mock steps:

Hermes Tasks panel updating at 4/10

### DIFF
--- a/agent/agent_init.py
+++ b/agent/agent_init.py
@@ -1790,9 +1790,10 @@ def init_agent(
     except Exception:
         pass
     
-    # In-memory todo list for task planning (one per agent/session)
-    from tools.todo_tool import TodoStore
-    agent._todo_store = TodoStore()
+    # Session-scoped todo list. The active copy stays in memory for fast tool
+    # calls while revisioned snapshots persist through the existing state DB.
+    from tools.todo_tool import create_session_todo_store
+    agent._todo_store = create_session_todo_store(agent)
     
     # Load config once for memory, skills, and compression sections
     try:
--- a/agent/turn_context.py
+++ b/agent/turn_context.py
@@ -854,6 +854,12 @@ def build_turn_context(
         else:
             with persist_lock:
                 agent._ensure_db_session()
+        # History hydration runs before the first-row create above. Persist its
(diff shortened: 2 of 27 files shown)


teknium1:
Salvaged and merged in #98248 (c0875ba503) — thank you @itsflownium! Your revision design, the dedicated `todo.updated` event, the resume-payload snapshot attach, and the desktop revision arbitration all landed under your authorship, along with @Adolanium's #97811 desktop merge fix which you had already incorporated.

One deliberate change: the `session_todo_state` table and TodoStore persistence hooks were dropped per maintainer direction (no parallel DB store for this). Canonical todo tool results already persist in conversation history, so resume paths derive the snapshot from the already-loaded transcript instead — same recovery behavior, zero schema changes and zero extra DB reads. Closing in favor of the merged salvage.

PR #96989, opened 2026-08-28
fix(gateway): arm loop-tick witness only on POSIX

Jackal991:
Closes #96956

## Summary

`loop_heartbeat_forever` in `gateway/shutdown_watchdog.py` called `asyncio.start_unix_server` unconditionally. That function only exists on POSIX (AF_UNIX asyncio support), so on native Windows every gateway start raised `AttributeError` — swallowed by the surrounding broad `except` into a noisy `WARNING` + traceback in the logs. The loop-tick liveness witness was never armed on Windows and the heartbeat payload always recorded `loop_tick_socket: false`.

That witness absence on Windows is deliberate fail-safe (documented in the adjacent comment block): a wedged native-Windows gateway keeps the graceful-drain backstop instead of an escalation verdict built on a witness that cannot exist. The bug was that the code attempted the impossible call anyway and "failed" into that state via an accident, logging a warning every single start. The stale-socket sweep immediately above the call site was already correctly gated with `if os.name == "posix":` — this PR extends that same gate to the server creation, mirroring it exactly, and logs the non-POSIX skip at DEBUG (expected condition, not a failure).

## Changes

- `gateway/shutdown_watchdog.py`: moved `tick_server = await asyncio.start_unix_server(...)` inside the existing `if os.name == "posix":` block; added an `else:` branch logging at DEBUG that the witness is intentionally not armed on non-POSIX. The outer `except Exception:` → `WARNING` path is untouched, so genuine POSIX bind failures (permissions, path length) still warn. Updated the comment block to describe the skip as explicit/gated rather than an `AttributeError` accident.
- `tests/gateway/test_shutdown_watchdog.py`: two regression tests:
  - `test_loop_tick_witness_skipped_intentionally_on_windows` — simulates a non-POSIX platform as seen from the module (the module's `os` binding is swapped for a proxy with `name = "nt"`; a direct global `os.name` patch would flip `pathlib.Path` dispatch and crash pytest itself) and installs a never-call recorder on `asyncio.start_unix_server`. Asserts the server creation is never attempted, no `Loop tick socket unavailable` warning is logged, and the payload still records `loop_tick_socket: false` (fail-safe preserved).
  - `test_loop_tick_witness_arms_on_posix` — the real POSIX path still arms the witness (`loop_tick_socket: true`).

## Tests

[...]

PR #91099, opened 2026-08-20
fix(gateway): supervise Windows watchdog exits

darrinmcgann:
## What does this PR do?

Fixes a Windows gateway reliability gap where Task Scheduler’s configured `RestartOnFailure` policy could not observe the actual gateway process.

The Scheduled Task runs a hidden VBS launcher through `wscript.exe`. That VBS previously called `WScript.Shell.Run(..., 0, False)`, detached from Python, and exited successfully immediately. If the gateway’s loop-liveness watchdog later terminated Python with exit code 75, Task Scheduler still saw only the already-successful VBS launcher and left the task `Ready` with no gateway process.

This change keeps the no-console `wscript.exe` design, but makes the VBS wait for Python and quit with Python’s exit code. The existing Task Scheduler `RestartOnFailure` policy can then observe code 75 and restart the task.

It also persists loop-liveness thread dumps to the existing watchdog log before hard exit. Previously those dumps were written only to uncaptured stderr under the hidden VBS launch, so the blocking frame was lost.

## Related Issue

Fixes #91097

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✅ Tests (adding or improving test coverage)

## Changes Made

- `hermes_cli/gateway_windows.py`
  - Keep the VBS launch hidden.
  - Wait for the Python gateway child.
  - Propagate the child exit code through `WScript.Quit`.
- `gateway/shutdown_watchdog.py`
  - Reuse the durable watchdog writer for loop-liveness exits.
  - Label liveness dumps separately and avoid duplicating the stderr dump.
- `tests/hermes_cli/test_gateway_windows.py`
  - Require hidden, supervised VBS behavior and exit-code propagation.
- `tests/gateway/test_loop_liveness_watchdog.py`
  - Require a durable liveness dump containing actual traceback content.

## How to Test

1. Run the focused tests:

   ```bash
   uv run --with pytest --with pyyaml python -m pytest \
     tests/hermes_cli/test_gateway_windows.py \
     tests/gateway/test_shutdown_watchdog.py \
     tests/gateway/test_loop_liveness_watchdog.py -q
   ```

2. Run focused lint:

   ```bash
   uv run --with ruff ruff check \
     hermes_cli/gateway_windows.py \
     gateway/shutdown_watchdog.py \
     tests/hermes_cli/test_gateway_windows.py \
     tests/gateway/test_loop_liveness_watchdog.py
   ```

3. On Windows, install/start the gateway task and confirm:
[...]

PR #97817, opened 2026-08-29
fix(gateway): handle missing unix server on Windows

up2itnow0822:
## Summary

Native Windows gateways no longer print an `AttributeError` traceback on every startup when `asyncio.start_unix_server` is unavailable. The loop heartbeat now takes the already-documented witness-disabled path explicitly and emits an informational message, while real bind failures on supported platforms still retain warning-level exception context. Regression coverage pins both the fail-safe `loop_tick_socket: false` payload and the absence of an exception warning.

## Test plan

- `scripts/run_tests.sh tests/gateway/test_shutdown_watchdog.py -q --tb=short` (`HERMES_PYTHON` pointed at the verified pytest-capable interpreter for the shallow clone)
- `scripts/run_tests.sh tests/gateway/test_loop_liveness_watchdog.py -q -k 'test_loop_scheduling_witness_is_served_by_the_loop_itself' --tb=short`
- `python -m py_compile gateway/shutdown_watchdog.py tests/gateway/test_shutdown_watchdog.py`
- Native Windows gateway restart: WhatsApp connected, heartbeat fresh, and no `start_unix_server` traceback

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
GPT_5.6_sol

### DIFF
--- a/gateway/shutdown_watchdog.py
+++ b/gateway/shutdown_watchdog.py
@@ -526,7 +526,7 @@ async def loop_heartbeat_forever(
     # disables the witness, and the payload flag tells probes that staleness is
     # no longer sufficient authority to escalate.
     #
-    # Windows: asyncio.start_unix_server raises (no AF_UNIX event-loop
+    # Windows: asyncio.start_unix_server is absent (no AF_UNIX event-loop
     # support), so the witness is PERMANENTLY absent there — the payload
     # records loop_tick_socket=False and every stale-file probe classifies
     # UNKNOWN, never WEDGED. That is deliberate fail-safe: a wedged native
@@ -569,9 +569,15 @@ async def loop_heartbeat_forever(
                 logger.debug(
                     "stale loop-tick socket sweep failed", exc_info=True
                 )
-        tick_server = await asyncio.start_unix_server(
-            _tick_socket_handler, path=str(tick_socket_path)
-        )
+        if not hasattr(asyncio, "start_unix_server"):
+            logger.info(
+                "Loop tick socket unsupported on this platform — liveness "
+                "probes will not escalate on a stale heartbeat"
+            )
+        else:
+            tick_server = await asyncio.start_unix_server(
+                _tick_socket_handler, path=str(tick_socket_path)
+            )
     except Exception:
         tick_server = None
         logger.warning(
--- a/tests/gateway/test_shutdown_watchdog.py
+++ b/tests/gateway/test_shutdown_watchdog.py
@@ -32,6 +32,49 @@ def test_resolve_shutdown_watchdog_delay_adds_grace():
     assert resolve_shutdown_watchdog_delay(10, grace_s=5) == 15.0
 
 
+def test_loop_heartbeat_missing_unix_server_is_clean_platform_fallback(
+    tmp_path, monkeypatch
+):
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed `gateway/shutdown_watchdog.py` (+ tests) for PR #97817. The intent — stop the noisy `AttributeError`/warning traceback on native Windows while keeping real bind failures diagnostic on supported platforms — is right, and the two regression tests correctly pin the no-warning path and the still-warns bind-failure path.

One evidence-based caveat worth verifying before relying on it:

- **`hasattr(asyncio, "start_unix_server")` only covers Python versions where the attribute is genuinely absent.** I verified on this Windows box with Python 3.14.7: `hasattr(asyncio, "start_unix_server")` is `False`, so the guard works there. But on older CPython versions (roughly ≤3.12) the module-level function is present and only fails at call time — `loop.start_unix_server(...)` raises `NotImplementedError` (and some loops raise `AttributeError`) when awaited. In that case `hasattr` returns `True`, the new branch is skipped, and the exact startup warning this PR intends to remove still fires on those Pythons.
- Suggested hardening (cheap): in addition to (or instead of) the `hasattr` check, treat `NotImplementedError` / `AttributeError` from the bind attempt as a platform-fail-safe too — e.g. broaden the `except Exception` around `start_unix_server` to detect `NotImplementedError` and log at `info` instead of `warning`. That makes the behavior version-independent rather than relying on the attribute being present/absent.

The tests themselves are good (the `monkeypatch.delattr(...)` and the failing-bind monkeypatch both exercise the branch logic deterministically). Non-blocking, but if the gateway targets the 3.11/3.12 range on Windows, the current guard may not achieve its stated goal on those versions.

monerostar:

Host: Windows 11 (build 26200), Python 3.11.15, pytest 9.1.1. Head `7fe543de63`.

### Live platform fact

On this interpreter:

```
hasattr(asyncio, "start_unix_server") = False
AttributeError: module 'asyncio' has no attribute 'start_unix_server'
```

Current `main` still does `await asyncio.start_unix_server(...)`. That hits the existing `except Exception` and logs a **WARNING with exc_info** (`AttributeError`) on every native Windows gateway start. Functionally the witness is already fail-safe (`loop_tick_socket` stays false); the noise is the bug.

The PR’s `hasattr` gate logs **INFO, no exception** on the same host. Bind failures on platforms that *have* the API still warn (covered by the second test).

### Tests

```
python -m pytest tests/gateway/test_shutdown_watchdog.py -o addopts= -q --tb=short
```

4 passed on this head, including:

- `test_loop_heartbeat_missing_unix_server_is_clean_platform_fallback`
- `test_loop_heartbeat_unix_server_bind_failure_still_warns`

No overlap with #91099 on the `start_unix_server` call. Small, correct, Windows-real.

teknium1:
Fixed on main via #100915, a salvage of #96989 (@Jackal991, submitted earlier on Aug 28) — merge commit 57d305d5. Same fix: `asyncio.start_unix_server` is now gated behind the existing `os.name == "posix"` check in `gateway/shutdown_watchdog.py`, so native Windows no longer logs the AttributeError traceback on every gateway start.

Thanks @up2itnow0822 for the independent fix — closing as a duplicate of the earlier submission.

PR #97818, opened 2026-08-29
fix(cron): honor cron.execution_retention instead of a hardcoded 1000-row cap

Navlem:
## What does this PR do?

`cron.execution_retention` was documented config that no code read. `cron/executions.py` hard-coded `MAX_TERMINAL_EXECUTIONS = 1000`, so setting the key changed nothing while `hermes config get cron.execution_retention` echoed the operator's value straight back.

The failure is quiet, and it corrupts audits rather than breaking anything loudly. On my fleet the key was set to `2500`, yet the ledger held **exactly 1000** terminal rows spanning 4.55 days at ~220 runs/day. A retention audit then compared that span against the assumed 2500 cap and concluded that history had been lost before the increase and the cap "was not full". The cap was full — at 1000. Any tooling reasoning about a requested audit window inherits that wrong conclusion.

The approach mirrors the existing sibling `_cron_output_keep()` in `cron/jobs.py` rather than inventing a new pattern:

- An explicit int in `MAX_TERMINAL_EXECUTIONS` still wins, so the existing `monkeypatch.setattr(executions, "MAX_TERMINAL_EXECUTIONS", 3)` test and any embedder override keep working. The default becomes `None`, meaning "resolve from config". The sentinel is `None` rather than a `!= _DEFAULT` comparison so that an override set to exactly `1000` is still honoured instead of silently falling through to config.
- Config is read via `load_config_readonly()` — the prune path only reads, so it skips the defensive deepcopy.
- Unreadable config, a non-numeric value, a non-mapping `cron` section, or a non-positive value falls back to the 1000 default. A non-positive cap would delete every terminal row on each write, destroying the audit trail the ledger exists to keep, so it is deliberately **not** honored as "prune everything". The module override remains available for a deliberate zero.

`cron.execution_retention` is also registered in `hermes_cli/config_defaults.py`. Without that entry `hermes config set cron.execution_retention` does not recognise the key, so the setting is unreachable through the supported path even once pruning reads it. The cron docs gain the row-cap-vs-day-count caveat.

## Related Issue

No existing issue — I searched open and closed issues and PRs for `execution_retention` and `MAX_TERMINAL_EXECUTIONS` and found none. Happy to file one first if you'd prefer that order.

## Type of Change

- [ ] 💥 Breaking change
- [ ] 📝 Documentation
- [ ] 🔧 Refactor

## How Has This Been Tested?

`tests/cron/test_execution_ledger.py`: **24 passed, 0 failed** via the canonical runner (`bash scripts/run_tests.sh`).

Five new tests cover config-driven capping, unusable values (missing / `0` / `-1` / non-numeric / `None` / non-mapping), the module override, an override equal to the default, and the defaults registration.

Two positive controls confirm the tests discriminate rather than merely passing:

| Variant under test | Result |
|---|---|
| This fix | 24 passed, 0 failed |
| Reverted to the hardcoded cap | **5 of the new tests fail** |
| `!= _DEFAULT` sentinel instead of `None` | **3 fail**, incl. `test_retention_override_equal_to_default_is_honoured` |

Full `tests/cron/`: **1058 passed, 3 failed**. The 3 failures are pre-existing and reproduce identically on unmodified `origin/main` in a clean `git worktree` — `TestScriptTimeoutTreeKill::test_cancel_path_also_tree_kills` plus two `test_script_claim_heartbeat.py` cancel tests, all blocked by the repo's own `tests/conftest.py` `_live_system_guard`, which refuses `os.kill`/`os.killpg` on PIDs outside the test process group. Unrelated to retention.

Live path confirmed: with `cron.execution_retention: 2500` configured, the resolver returns `2500` where it previously returned `1000`.

## Note

Supersedes #98362, which fixed the same bug independently. Neither version was complete alone: #98362's `!= _DEFAULT` sentinel silently ignores an override equal to the default, while this branch's original form omitted the `config_defaults.py` registration. This PR combines both halves.

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes
- [x] I have made corresponding changes to the documentation

### DIFF
--- a/cron/executions.py
+++ b/cron/executions.py
@@ -21,7 +21,14 @@ from hermes_time import now as _hermes_now
 # dashboard operations that temporarily enter another profile cannot leak that
 # profile's execution records into the import-time home.
 EXECUTIONS_FILE: Optional[Path] = None
-MAX_TERMINAL_EXECUTIONS = 1000
+
+# Terminal-row cap. ``None`` resolves from ``cron.execution_retention`` at prune
+# time (mirrors ``cron.output_retention`` in cron/jobs.py), so a config change
+# takes effect without a reinstall. Assigning an int overrides config outright —
+# the sentinel is ``None`` rather than the default value so that a test or
+# embedder setting the override to exactly 1000 is still honoured.
+_DEFAULT_MAX_TERMINAL_EXECUTIONS = 1000
+MAX_TERMINAL_EXECUTIONS: Optional[int] = None
 _TERMINAL_STATES = ("completed", "failed", "unknown")
 _lock = threading.RLock()
 _PROCESS_ID = uuid.uuid4().hex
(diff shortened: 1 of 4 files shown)


Enough1122:

Reviewed `cron/executions.py` (+ tests + docs) for PR #97818. The fix is correct and closes a real silent failure: `cron.execution_retention` was echoed by `hermes config get` but never read by the pruner. `_max_terminal_executions()` resolving config with a safe fallback (non-positive / non-numeric → default 1000, never "unlimited") is the right behavior, and the test matrix covers missing, zero, negative, and non-numeric values. The docs update correctly notes it's a row cap, not a day count.

Non-blocking observations:

1. **The cap is re-resolved from config on every prune/write path** (`_max_terminal_executions()` → `load_config_readonly()` inside `_prune_unlocked`). If `load_config_readonly()` re-reads YAML on each call, busy fleets with many terminal writes pay a config parse per write. If it's cached/memoized this is a non-issue — worth confirming. (Could be hoisted to resolve once per ledger open if not.)

2. **Edge:** if a user sets `cron.execution_retention` to a `dict`-shaped or otherwise non-int value, `int(...)` raises and falls back to the default silently. That's the documented fallback contract, so acceptable — but it also means a typo'd config value is silently ignored rather than surfaced. The existing `except Exception → default` is consistent with how `cron.output_retention` is handled; no change needed, just noting the silent-fallback posture is now shared by both.

3. **Minor:** the module-level `MAX_TERMINAL_EXECUTIONS` changed from an `int` constant to `Optional[int] = None`. Any external code importing the old constant expecting an `int` would now get `None`; the tests handle this via monkeypatch. Worth a quick grep that no other module reads it as an int.

The change is minimal, well-tested, and the documentation accurately sets expectations. Nothing blocking.

PR #97828, opened 2026-08-29
test: fix mock type errors and sync call handling in conversation loop edge cases

forumevi:
### Summary
Fixes `TypeError` issues caused by unconfigured mock attributes (`_skill_nudge_interval`, `_iters_since_skill`) and updates `run_conversation` call handling to synchronous in unit tests.

### Key Changes
- Updated `mock_agent` fixture in `tests/test_conversation_loop_edge_cases.py` to correctly initialize integer nudge intervals and session counters.
- Adjusted test assertions to sync execution for `run_conversation`.

### Testing
- `python -m pytest tests/test_conversation_loop_edge_cases.py` (3 passed)

### DIFF
--- a/tests/test_conversation_loop_edge_cases.py
+++ b/tests/test_conversation_loop_edge_cases.py
@@ -1,113 +1,57 @@
 import pytest
-from unittest.mock import MagicMock, patch
-
+from unittest.mock import MagicMock
+from agent import conversation_loop
 
 @pytest.fixture
 def mock_agent():
-    """Ajan istemcisinin tip, string ve sayısal durum hatalarını önleyen eksiksiz mock yapısı."""
+    """
+    Fixture to initialize a properly configured mock agent object.
+    Ensures numerical session and iteration counters are set to prevent TypeError.
+    """
     agent = MagicMock()
-    # Düz dict dönen sync/mock yanıt
-    agent.run.return_value = {"status": "success", "response": "OK"}
-    
-    # Sayısal ve mantıksal varsayılan değerler
-    agent._memory_nudge_interval = 0
-    agent._user_turn_count = 0
-    agent._turns_since_memory = 0
-    agent._skill_nudge_interval = 0
+    agent._skill_nudge_interval = 5
     agent._iters_since_skill = 0
-    agent.max_iterations = 10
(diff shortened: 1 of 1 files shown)


Enough1122:

Reviewed `tests/test_conversation_loop_edge_cases.py` for PR #97828. The fixture correctly initializes the numeric session/iteration counters (`_skill_nudge_interval`, `_iters_since_skill`, etc.) that previously left MagicMock unconfigured, and the tests target `run_conversation`'s `moa_config` / `stream_callback=None` edge handling. I confirmed `run_conversation` in `agent/conversation_loop.py` is a plain (synchronous) `def`, so calling it synchronously in the tests is valid.

Non-blocking observations:

1. **The assertions are weak — `assert result is not None`.** Because `run_conversation` is synchronous, the code path does execute, but "not None" would pass for a wide range of wrong behavior. If the intent is to pin "invalid moa_config does not crash and returns a usable result", assert something stronger about the result shape (e.g. a key like `final_response` or a status field) rather than only non-None.

2. **The first test's try/except pattern masks failures:** `except Exception as e: pytest.fail(...)` re-raises as a failure, which is fine, but the `for config in invalid_configs` loop means a crash in an earlier iteration still reports the later iteration correctly — acceptable. More notably, if `run_conversation` returns a coroutine-free dict but the mocked `agent` doesn't actually exercise the moa_config branch (the fixture's `_ra` patch replaces `agent.conversation_loop._ra`), the test may pass without exercising the real invalid-config path. Worth confirming the mock exercises the branch that consumes `moa_config`.

3. **Style:** the docstrings/comments are in Turkish while the rest of the repo uses English — minor, but a pass to English would match conventions. Also the file ends without a trailing newline and has trailing whitespace on a couple of lines.

4. **`test_run_conversation_type_hints_importability`** asserts `"run_conversation" in __all__` — reasonable, though it only verifies the export, not behavior.

The fixture is a genuine improvement for mocking the agent surface; the tests just assert a bit loosely. Non-blocking.

forumevi:
Updated! I've refactored tests/test_conversation_loop_edge_cases.py to strengthen assertions, clean up whitespace, and translate all docstrings to English per the review feedback. Ready for merge! 🚀

PR #94838, opened 2026-08-25
fix(browser): actionable error + Xvfb fallback for headed + no DISPLAY

Finn763:
## What Problem This Solves

Closes #94827.

When `browser.headed: true` is set in `config.yaml` and `DISPLAY` /
`WAYLAND_DISPLAY` are not visible to the spawned Chromium (either because the
profile `.env` was not loaded into the agent-browser subprocess, or because the
host genuinely has no display server), the browser still crashes immediately:

```
Missing X server or $DISPLAY
The platform failed to initialize. Exiting.
```

The user gets a raw Chromium error with no hint about what to do. On Linux
servers (the common case for headed browser work), the only escape was to
manually attach a Chromium via `browser.cdp_url` with a per-restart UUID.

## What This PR Changes

Adds a layered mitigation in `tools/browser_tool.py` around `_run_browser_command`:

1. **Pre-launch auto-fallback (Linux only).** When `browser.headed: true` and no
   `DISPLAY` / `WAYLAND_DISPLAY` / `XAUTHORITY` is visible to Hermes, attempt to
   spin up an `Xvfb` virtual display in a detached process group. If Xvfb is on
   PATH and binds the socket, the new `DISPLAY=:NN` is merged into the
   agent-browser subprocess env right before Popen, so the spawned Chromium
   inherits a usable display. If Xvfb is missing or fails, we fall through to
   the post-mortem error path below — never raises.
2. **Post-mortem actionable error.** If the browser subprocess still exits
   non-zero with a stderr matching Chromium's "no display server" signature
   (`Missing X server or $DISPLAY` or `The platform failed to initialize.
   Exiting.`), the previous raw stderr is replaced with a copy that:
   - Names the env vars that unblock headed mode (`DISPLAY`, `WAYLAND_DISPLAY`).
   - Points the operator at the auto-fallback (`apt install xvfb` /
     `pacman -S xorg-server-xvfb`) and the profile `.env` they should set.
   - Explains the headed-vs-headless distinction so the user can choose to
     flip `browser.headed: false` if they do not actually need a window.

The auto-fallback is opt-in via the existing `browser.headed` flag — if the
user asked for headed, we make it work; if they already have `DISPLAY=:0`
exported, we never touch anything.

## Evidence

Regression tests in `tests/tools/test_browser_headed_mode.py`:

- `TestMissingDisplayDetection::test_detects_missing_x_server_pattern` —
  the detector matches both Chromium banners and ignores unrelated stderr.
[...]

Issue #94827, opened 2026-08-25
headed: true browser crashes with Missing X server despite DISPLAY in profile .env


BlackKite84:
## Description

When `browser.headed: true` is set in `config.yaml` and `DISPLAY=:0` / `WAYLAND_DISPLAY=wayland-0` are configured in the profile `.env` file, Hermes still spawns a headless Chromium that crashes immediately:

```
Missing X server or $DISPLAY
The platform failed to initialize. Exiting.
```

The browser subprocess does not inherit environment variables from the profile `.env`, even though they are set there.

## Workaround

Start Chromium externally (headed, with correct Wayland flags) and attach via `browser.cdp_url`:

```bash
hermes config set browser.cdp_url "ws://127.0.0.1:9222/devtools/browser/<UUID>"
```

This works but the UUID changes on every Chromium restart.

## Expected Behavior

When `browser.headed: true` is set, Hermes should either:

1. **Inherit environment variables** from the profile `.env` (or `env_passthrough` list) for the browser subprocess, so that `$DISPLAY` and `$WAYLAND_DISPLAY` are available to the spawned Chromium.
2. Or provide a dedicated config key like `browser.env` to pass environment variables to the browser process.

## Environment

- OS: Linux (Arch-based CachyOS)
- Display: KDE Plasma on Wayland, Xwayland on :0
- Hermes: latest, profile-based setup
- Config: `browser.headed: true`, `browser.backend: agent_browser`, `browser.cdp_url: ""`

## Steps to Reproduce

1. Set `DISPLAY=:0` and `WAYLAND_DISPLAY=wayland-0` in `~/.hermes/profiles/<name>/.env`
2. Set `browser.headed: true` in `config.yaml`
3. Call `browser_navigate(url="http://example.com")`
4. Observe: Chromium exits with "Missing X server or $DISPLAY"

Finn763:
Taking this on. Headed:true browser launch crashes with Missing X server despite DISPLAY — will trace the headed-mode launcher, surface an actionable 'no DISPLAY + no Xvfb fallback' error instead of a raw browser crash, and add an auto-fallback path (Xvfb on Linux when headed + no DISPLAY). Regression tests for both error surface and Xvfb fallback. Push fork with literal PR headings + Closes #94827.

PR #97830, opened 2026-08-29
fix(desktop): explain a dead local $DISPLAY instead of crashing silently

burlito:
## What does this PR do?

`hermes desktop` can crash with a cryptic native error and no way for app code to intercept or explain it:

```
[hermes] remote display detected (ssh-session); disabling GPU hardware acceleration to prevent flicker
[hermes] using password-store backend: kwallet6
[hermes] install stamp: ...
[...:ERROR:ui/ozone/platform/x11/ozone_platform_x11.cc:256] Missing X server or $DISPLAY
[...:ERROR:ui/aura/env.cc:257] The platform failed to initialize.  Exiting.
```

This happens when `$DISPLAY` is set to a locally-numbered display (`:N`) that has no backing socket in `/tmp/.X11-unix`. The most common way to hit it: a long-lived tmux/screen pane outlives the X session that originally set `$DISPLAY` — an Xwayland restart, or an SSH X11 forward reconnecting, hands out a new display number, but panes already open keep the stale value. Every `hermes desktop` invocation from that pane then hits the same opaque crash with no hint about what's actually wrong or how to fix it.

This PR adds `detectDeadLocalDisplay()` (in the same style/location as the existing `detectRemoteDisplay()`) to catch this specific, checkable case before Electron gets far enough to crash, and print an actionable message instead:

```
[hermes] DISPLAY=:11 set, but /tmp/.X11-unix/X11 doesn't exist — Hermes can't open a window here.
This usually means $DISPLAY was inherited from a session that's gone (a tmux/screen pane kept an
old display number across an Xwayland restart or SSH reconnect). Open a fresh shell and try again.
```

Forwarded displays (`host:N`, e.g. classic SSH X11 forwarding over TCP) aren't covered — verifying those needs a network round trip, which doesn't fit the synchronous, pre-`app-ready` check this runs as. Those still fall through to Electron's existing (unfriendly) failure.

Linux-only by construction (`detectDeadLocalDisplay` returns `null` immediately on any other `platform`), so no macOS/Windows impact.

## Related Issue

No existing issue covered this exact failure mode — I searched open/closed issues and PRs first (#94827 / #94838 is a related-looking but distinct symptom in the `browser` headed-mode tool, not this one).

Fixes #

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `apps/desktop/electron/bootstrap-platform.ts`: add `detectDeadLocalDisplay()`, exported alongside `detectRemoteDisplay()`
- `apps/desktop/electron/main.ts`: call it before the existing remote-display/GPU check; print an actionable message and `process.exit(1)` instead of letting Electron crash
- `apps/desktop/electron/bootstrap-platform.test.ts`: 3 new unit tests covering the dead-socket case, the healthy-socket case, and the displays it deliberately leaves alone (forwarded, Wayland, unset, non-Linux)
- `contributors/emails/sasha@pokite.me`: contributor-attribution mapping (required by `contributor-check.yml` for this commit's author email)

## How to Test

1. `cd apps/desktop && npx vitest run --project electron electron/bootstrap-platform.test.ts` — 16/16 pass (13 existing + 3 new)
2. Manual repro: `export DISPLAY=:99` (any display number with no live X server), run `hermes desktop` — before this change it crashes with the raw Ozone error; after, it prints the actionable message above and exits 1
3. `npx tsc -p tsconfig.electron.json --noEmit` — clean
4. `eslint electron/bootstrap-platform.ts electron/bootstrap-platform.test.ts electron/main.ts` — clean

### Code

- [x] My PR contains **only** changes related to this fix (plus the contributor-mapping file the CI attribution check requires)
- [x] I've run `pytest tests/ -q` and all tests pass — N/A, this change touches only `apps/desktop` (TypeScript/Electron); no Python files changed. Ran the equivalent desktop suite instead (see "How to Test").
- [x] I've tested on my platform: Linux (Fedora), physical Xwayland session — the exact scenario this fixes

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A, no user-facing docs/config affected
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — N/A, no config keys added
- [x] I've updated `CONTRIBUTING.md` or `AGENTS.md` if I changed architecture or workflows — N/A, no architecture/workflow change
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — Linux-only check, gated on `platform === 'linux'`, no effect elsewhere
- [x] I've updated tool descriptions/schemas if I changed tool behavior — N/A, not a tool

## Screenshots / Logs

Before/after console output included above under "What does this PR do?" / "How to Test".


monerostar (commented):
Ubuntu 26.04 on a local 5800X box (kernel 7.0.0-30-generic, Xwayland). Live sockets: /tmp/.X11-unix/X0 and X1 exist. X11 and X99 do not.

Imported detectDeadLocalDisplay from this PR and called it against real fs.existsSync:

- DISPLAY=:0 or :1, no WAYLAND_DISPLAY -> null (healthy sockets)
- DISPLAY=:11 -> DISPLAY=:11 set, but /tmp/.X11-unix/X11 doesn't exist
- DISPLAY=:99 -> same shape for X99
- WAYLAND_DISPLAY=wayland-0 plus dead DISPLAY=:11 -> null (skip, as written)
- forwarded localhost:10.0, unset DISPLAY, darwin -> null

This session's normal env is DISPLAY=:0 plus WAYLAND_DISPLAY=wayland-0, so the default process.env path also returns null. That matches the Wayland skip.

vitest did not start on this install (missing @rolldown/plugin-babel). The function itself matches the new cases on this Linux box.

Looks good.

PR #42697, opened 2026-06-09
fix(skills_sync): refresh stale manifest when user copy matches bundled

liuhao1024:
## What does this PR do?

Fixes a logic flaw in `sync_skills()` where curator-archived bundled skills were falsely flagged as "user-modified" on every subsequent sync, permanently blocking updates.

## Related Issue

Fixes #42682

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `tools/skills_sync.py`: When `user_hash != origin_hash`, additionally check if `user_hash == bundled_hash`. If the user copy matches the current bundled version, refresh the stale manifest instead of flagging as user-modified. This covers the curator-archived skill scenario where the bundled version was updated but the manifest wasn't synced.
- `tests/tools/test_skills_sync.py`: Added `test_stale_manifest_synced_when_user_matches_bundled` covering the stale-manifest scenario. Updated `test_reset_clears_stuck_user_modified_flag` to reflect that sync now correctly handles this case.

## How to Test

1. Run `pytest tests/tools/test_skills_sync.py -x -q` — all 57 tests pass
2. The new test `test_stale_manifest_synced_when_user_matches_bundled` specifically validates: when user copy matches bundled but manifest origin is stale, sync refreshes the manifest instead of flagging as user-modified
3. The updated `test_reset_clears_stuck_user_modified_flag` confirms reset still works as a manual override even after sync handles the case

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
- [x] My PR contains **only** changes related to this fix/feature (no unrelated commits)
- [x] I've run `pytest tests/tools/test_skills_sync.py -q` and all tests pass
- [x] I've added tests for my changes (required for bug fixes, strongly encouraged for features)
- [x] I've tested on my platform: macOS

### Documentation & Housekeeping

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — or N/A
- [x] I've updated `CONTRIBUTING.md` or `AGENTS.md` if I changed architecture or workflows — or N/A
[...]

PR #97834, opened 2026-08-29
fix(skills): verify destination integrity before the unchanged-source skip

### LINKED ISSUE #97791: skills: sync short-circuits on source hash, so destination drift is never detected (skills silently frozen)
## Bug Description

`sync_skills()` decides whether a bundled skill needs updating by comparing the **bundled source** hash against the hash recorded in `.bundled_manifest`. When those match it `continue`s without ever hashing the **destination**. Any drift on the destination side is therefore invisible, and the sync reports `up to date` / `skipped` forever — until upstream happens to touch that skill again.

`tools/skills_sync.py:886-892`:

```python
# If the bundled source still matches the version recorded when
# it was installed, there is no update to apply. Avoid recursively
# hashing the user's copy just to rediscover that fact; when the
# bundled source changes, the normal user-modification check below
# still protects local edits before any overwrite.
if origin_hash and bundled_hash == origin_hash:
    skipped += 1
    continue
```

The comment is only true if nothing else can modify the destination. In practice several things can: the skill curator (`prune_builtins`, `.curator_backups` restores), `repair-official --restore` backups, partial installs, a user restoring an older copy by hand, or an interrupted sync.

Once the destination diverges while the source is unchanged, the skill is **permanently frozen** and no command reports a problem.

## Steps to Reproduce

```bash
# 1. Start from a synced profile
hermes update                     # -> "Skills are up to date"

# 2. Make the destination drift without touching the bundled source
#    (simulates a curator restore / partial install / older copy)
rm -rf ~/.hermes/skills/productivity/pdf/references
rm -f  ~/.hermes/skills/productivity/pdf/LICENSE

# 3. Sync again
hermes update
```

## Expected Behavior

The sync notices the destination no longer matches what the manifest says was installed, and either restores the bundled files or reports the skill as needing repair.

## Actual Behavior

Reports `✓ Skills are up to date` / `skipped`. `hermes skills list-modified` does not list it either, because it was never classified as modified. The skill stays broken indefinitely and silently.

## Real-world impact (how this was found)

Found on a 7-profile install after updating v0.20.5 → v0.20.6. Four bundled skills had silently drifted, all reported as in-sync:

| Skill | Actual on-disk state | Manifest said |
|---|---|---|
| `pdf` | old flat layout (`reference.md`, `forms.md`, `LICENSE.txt`) | current hash |
| `docx` | missing `references/`, `LICENSE`, several `scripts/` | current hash |
| `xlsx` | missing `references/`, `LICENSE`, several `scripts/` | current hash |
| `python-debugpy` | `scripts/` absent entirely | current hash |

Verified directly — for each of these, `manifest[name] == _dir_hash(bundled_src)` while `_dir_hash(dest) != _dir_hash(bundled_src)`, which is exactly the short-circuit above:

```
skill                manifest(origin)  bundled(src)  dest(actual)   verdict
docx                 62fa1308…         468e9570…     62fa1308…      SKIPPED (src==manifest) ** DRIFT MASKED **
pdf                  15ce4e3f…         934abf08…     15ce4e3f…      SKIPPED (src==manifest) ** DRIFT MASKED **
xlsx                 9ebebf27…         0dad9c69…     9ebebf27…      SKIPPED (src==manifest) ** DRIFT MASKED **
python-debugpy       83cd8743…         76a10139…     83cd8743…      SKIPPED (src==manifest) ** DRIFT MASKED **
```

## Related second-order bug: a re-synced skill stays flagged forever

`_is_tracked_user_modification(origin_hash, user_hash)` (`tools/skills_sync.py:907`) compares the user copy against the **stale** `origin_hash` only. After repairing a skill so that it is byte-identical to the *current* bundled version, it is still reported as `user-modified` and skipped, because `user_hash != origin_hash` even though `user_hash == bundled_hash`.

Observed after repairing `hermes-agent` to exactly the bundled content:

```
creative   sync=ok  copied=0 updated=0 user_mod=['google-workspace', 'computer-use', 'hermes-agent']
```

`hermes skills reset <name>` does break the loop (as its docstring says), but the classification could self-heal: if `user_hash == bundled_hash`, the copy is stock — re-baseline the manifest instead of reporting a modification.

## Suggested fix

In the short-circuit, treat the manifest as a claim to be verified rather than the truth. Options, cheapest first:

1. Keep the fast path but make it cheap-verifiable — record a file count / mtime-max / total size alongside the hash and re-check that before skipping; only fall back to a full `_dir_hash(dest)` when the cheap check disagrees.
2. Verify destination existence and completeness (all manifest-known paths present) even when hashes match.
3. Add an explicit `hermes skills verify [--repair]` that always hashes destinations, so drift is at least discoverable without a full re-hash on every sync.

And in the classifier, add `if user_hash == bundled_hash: re-baseline; skipped += 1; continue` ahead of the `_is_tracked_user_modification` check.

## Environment

- Hermes v0.20.6, `main @ aff5125f8e`
- 7 profiles (default + creative, design, dev, finance, general, research)
- Skills are symlink farms into a shared `~/.hermes/skill-store/`, so a single store copy backs all profiles (not required to reproduce, but it is how the drift went unnoticed across every profile at once)

liuhao1024:
## What does this PR do?

Fixes the destination-drift blind spot in `sync_skills()` reported in #97791. The unchanged-source fast path (`origin_hash == bundled_hash` → skip) treats the manifest hash as the truth and never looks at the destination, so any drift introduced while the bundled source was unchanged — a curator `prune_builtins`/`.curator_backups` restore, `repair-official --restore` leftovers, a partial install, an interrupted sync, or a hand-restored older copy — stays masked as "up to date" forever. On the reporter's real 7-profile install, four bundled skills (`pdf`, `docx`, `xlsx`, `python-debugpy`) had silently drifted (missing `references/`, `LICENSE`, `scripts/`) while every sync reported them in sync.

The fix keeps the fast path but makes it cheap-verifiable (option 1 in the issue): the skip now requires a stat fingerprint match (file count + total size, no content reads) between the destination and the bundled source. A disagreement falls through to the full content hash and the existing classification path, so a drifted copy is reported as `user-modified` — exactly the documented contract for a changed copy ("treated as user-modified and skipped forever, so your edits never get stomped"), which today is silently bypassed for this case. Nothing is ever overwritten; `hermes skills reset <name>` remains the documented repair path.

**Scope boundary (deliberately excluded):** this PR does *not* change the documented reset-based recovery contract for the issue's second-order case — a destination byte-identical to the current bundled version but recorded under a stale origin hash still reports as user-modified, as documented in `website/docs/user-guide/features/skills.md` ("`hermes skills reset` is the escape hatch"). Changing that policy was already reviewed on #42697, where the sweeper asked for maintainer agreement first; auto-re-baselining in sync would also bypass the protection that makes a manual restore deliberate. Only the *visibility* of destination drift is fixed here.

## Related Issue

Fixes #97791

## Type of Change

## Changes Made

- `tools/skills_sync.py`: added `_stat_fingerprint()` — a cheap (file count, cumulative size) tree fingerprint that never reads file contents; returns `None` on scan errors so callers fail toward the full hash.
- `tools/skills_sync.py`: the unchanged-source fast path now skips only when the destination's stat fingerprint matches the bundled source's; a drifted destination falls through to the full content hash and the existing `_is_tracked_user_modification()` classification, where it is reported as `user-modified` (visible instead of masked). Healthy copies keep the fast path — no content hashing is added to the common case.
- `tests/tools/test_skills_sync.py`: two regression tests — a drifted destination is reported (not masked), and a second sync of an already-synced tree stays fully skipped (fast-path preservation).

## How to Test

1. `pytest tests/tools/test_skills_sync.py -q` → **36 passed**. Observed result: all green, including the drift regression test.
2. Red/green check against the pre-fix code: `git checkout upstream/main -- tools/skills_sync.py && pytest tests/tools/test_skills_sync.py -q` → **1 failed** (`test_drifted_destination_reported_not_masked`), proving the test pins the bug; restoring the fix returns to 36 passed.
3. Manual repro of the issue's Steps to Reproduce: sync once (manifest baselined), delete `main.py` from the installed copy, sync again → the skill is now reported `~ <name> (user-modified, skipping)` instead of `✓ Skills are up to date`, so the documented `hermes skills reset <name>` path can repair it.

Test scope note: I also ran the neighboring skills suites and the `test_cmd_update.py`/`test_profiles.py` consumers of `sync_skills`. `test_cmd_update.py` shows 8 pre-existing failures on this machine that reproduce identically on a clean `upstream/main` checkout (local environment interference, unrelated to this diff — none touch `tools/skills_sync.py` behavior assertions).

### Code

## For New Skills

N/A

### DIFF
--- a/tests/tools/test_skills_sync.py
+++ b/tests/tools/test_skills_sync.py
@@ -547,40 +547,6 @@ class TestSyncSkills:
         )
         assert "new-skill" not in result["updated"]
 
-    def test_stock_copy_with_stale_origin_hash_is_rebaselined(self, tmp_path):
-        """A destination byte-identical to the current bundled version but
-        recorded under a stale origin hash must self-heal (re-baseline),
-        not be reported as user-modified on every sync forever (#97791)."""
-        bundled = self._setup_bundled(tmp_path)
-        skills_dir = tmp_path / "user_skills"
-        manifest_file = skills_dir / ".bundled_manifest"
-
-        # Simulate: skill was installed from an older bundled version...
-        user_skill = skills_dir / "old-skill"
-        user_skill.mkdir(parents=True)
-        (user_skill / "SKILL.md").write_text("# Old v1")
-        manifest_file.write_text(f"old-skill:{_dir_hash(user_skill)}\n")
-
-        # ...then the bundled version changed and the user repaired their
-        # copy to be byte-identical to the new bundled content.
-        (bundled / "old-skill" / "SKILL.md").write_text("# Old")
-        shutil.rmtree(user_skill)
-        shutil.copytree(bundled / "old-skill", user_skill)
-        assert _dir_hash(user_skill) == _dir_hash(bundled / "old-skill")
-
-        with self._patches(bundled, skills_dir, manifest_file):
-            result = sync_skills(quiet=True)
-            manifest = _read_manifest()
-
-        assert "old-skill" not in result["user_modified"], (
-            "A stock copy identical to the current bundled version is still "
-            "flagged as user-modified forever"
-        )
-        assert "old-skill" not in result["updated"]
-        assert manifest["old-skill"] == _dir_hash(bundled / "old-skill"), (
-            "Manifest was not re-baselined to the current bundled hash"
-        )
-
     def test_second_sync_of_stock_copy_stays_skipped(self, tmp_path):
         """The unchanged-source fast path must keep skipping a healthy copy:
         no update churn, no user-modified noise on an already-synced tree."""
@@ -652,12 +618,11 @@ class TestResetBundledSkill:
         manifest_file.write_text("google-workspace:STALEHASH000000000000000000000000\n")
 
         with self._patches(bundled, skills_dir, manifest_file):
-            # Sanity check: since the #97791 re-baseline fix, sync self-heals
-            # exactly this state (a stock copy under a stale origin hash) and
-            # no longer flags it user_modified forever; reset stays as the
-            # escape hatch for genuinely customized copies.
+            # Sanity check: without reset, sync would flag it user_modified
+            # (the documented reset-based recovery contract — see
+            # website/docs/user-guide/features/skills.md).
             pre = sync_skills(quiet=True)
-            assert "google-workspace" not in pre["user_modified"]
+            assert "google-workspace" in pre["user_modified"]
 
             # Reset (no --restore) should clear the manifest entry and re-baseline
             result = reset_bundled_skill("google-workspace", restore=False)
--- a/tests/tools/test_skills_sync.py
+++ b/tests/tools/test_skills_sync.py
@@ -521,6 +521,84 @@ class TestSyncSkills:
             assert "new-skill" in result2["copied"]
             assert (skills_dir / "category" / "new-skill" / "SKILL.md").exists()
 
+    def test_drifted_destination_reported_not_masked(self, tmp_path):
+        """A destination that drifted while the bundled source was unchanged
+        (curator restore, partial install, interrupted sync) must surface as
+        user-modified, not stay silently "up to date" forever (#97791)."""
+        bundled = self._setup_bundled(tmp_path)
+        skills_dir = tmp_path / "user_skills"
+        manifest_file = skills_dir / ".bundled_manifest"
+
+        # First sync: everything copied, manifest baselined.
+        with self._patches(bundled, skills_dir, manifest_file):
+            sync_skills(quiet=True)
+            manifest_after_first = _read_manifest()
+        assert manifest_after_first["new-skill"] == _dir_hash(bundled / "category" / "new-skill")
+
+        # Drift the destination without touching the bundled source.
+        (skills_dir / "category" / "new-skill" / "main.py").unlink()
+
+        with self._patches(bundled, skills_dir, manifest_file):
+            result = sync_skills(quiet=True)
+
+        assert result["user_modified"] == ["new-skill"], (
+            "A drifted destination was masked by the manifest hash match — "
+            "the skill stays broken indefinitely and no command reports it"
+        )
+        assert "new-skill" not in result["updated"]
+
+    def test_stock_copy_with_stale_origin_hash_is_rebaselined(self, tmp_path):
+        """A destination byte-identical to the current bundled version but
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed `tools/skills_sync.py` (+ tests) for PR #97834. The fast-path fix is exactly right: the unchanged-source skip (`bundled_hash == origin_hash`) now gated on a cheap `_stat_fingerprint` (file count + total size) so destination drift that never touches the bundled source surfaces instead of being masked as "up to date" forever (#97791). The tests pin both directions — drifted destination flagged as `user_modified`, and a healthy second sync still skipped with no noise.

Non-blocking observations:

1. **Competing/overlapping fix — #97840 addresses the same #97791 in the same file.** Both introduce a near-identical stat helper (`_stat_fingerprint` here vs `_dir_file_stats` there) and both restructure the same fast-path block in `sync_skills`. Whichever merges first will likely conflict with the other; the maintainers should reconcile them (pick one helper + one self-heal strategy, or fold #97840's stale-manifest re-baseline in on top of this one) rather than merge both as-is.

2. **A stat match still skips without a content hash — correct for perf, with a known trade-off.** Two trees with identical file count+size but different content (same-size content drift) would still be treated as "up to date". #97840's variant always hashes the destination (no fast-path savings); this PR keeps the fast path but accepts the same-size-drift blind spot. Given the bug being fixed is about *missing/moved files* (structural drift), the count+size fingerprint covers the reported class; a comment noting the same-size limitation would help.

3. **Minor:** `_stat_fingerprint` returns `None` on scan error, and the caller treats that as "fingerprints disagree" (safe fallback to full hash) — good defensive choice. The `rglob` + `stat` cost is bounded and runs only when the manifest hash matches.

The fix is minimal, well-tested, and preserves the perf intent of the fast path. The only real action item is coordinating with #97840.

liuhao1024:
Thanks for the review — and for flagging #97840. Coordination context for the maintainers: this PR went up ~12 minutes before #97840 and the two do overlap on the same fast-path block. Where they differ: this one keeps a cheap stat fingerprint in front of the manifest-hash skip (preserving the perf fast path, with the known same-size-drift blind spot called out above) and treats scan errors as disagree (full-hash fallback); #97840 hashes the destination on every sync and adds a stale-manifest re-baseline. Happy to rebase onto whichever one is preferred — and the same-size limitation deserves a one-line comment either way, which I'll add here shortly.

PR #97835, opened 2026-08-29
fix(profiles): make_targz writes to a temp file and renames, not the destination directly

nftpoetrist:
## Summary

`make_targz()` (`hermes_cli/archive_safe.py`) — the shared writer behind both `hermes profile export` and `hermes kanban export` — opened the destination `.tar.gz` path directly with `tarfile.open(archive_path, "w:gz")`, which truncates the file the instant it opens. If `tf.add()` fails partway through (disk full, permission revoked mid-write, process killed), whatever was previously at that path is destroyed and replaced with a truncated/corrupt archive.

This is the exact same failure shape `a7e7de6407` fixed for the desktop gateway file-save path — one commit earlier in the same merge window — but the fix was never propagated to this shared archive-writing primitive, even though board export gained a brand-new caller into it (`export_board`) in that same window. The existing `export_profile` caller carries the same bug and predates the refactor.

Reproduced empirically against a pre-existing file at the destination path:
```
before: 38 bytes
after (current buggy behavior): 28 bytes
content survives? False
```

## Fix

`make_targz()` now writes into a sibling temp file created with `tempfile.mkstemp()` in the same directory as the destination (so the final step is a same-volume, atomic `os.replace()`), and only replaces the destination after the archive is fully written and the file descriptor is closed. The temp file is unlinked on any failure. This mirrors the mkstemp + os.replace pattern already established elsewhere in this codebase (`agent/secret_sources/_cache.py`, `cron/jobs.py`, `gateway/status.py`, `gateway/rich_sent_store.py`).

## Test plan

- [x] New regression test (`tests/hermes_cli/test_archive_safe.py`) reproduces the bug via a monkeypatched `tarfile.TarFile.add` failure mid-write and asserts a pre-existing archive survives untouched, with no leftover temp file.
- [x] Mutation-verified: stashed the fix, confirmed the new test fails against the original code (sentinel content replaced by a truncated gzip header), restored the fix, confirmed it passes.
- [x] Added round-trip and overwrite-on-success tests to confirm normal behavior is unchanged.
- [x] Full `tests/hermes_cli/test_kanban_transfer.py` (22 tests) and `tests/hermes_cli/test_profile_export_credentials.py` + `tests/hermes_cli/test_profiles.py` (60 tests, 2 pre-existing skips) pass unchanged.

### DIFF
--- a/hermes_cli/archive_safe.py
+++ b/hermes_cli/archive_safe.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 import os
 import shutil
 import tarfile
+import tempfile
 from pathlib import Path, PurePosixPath, PureWindowsPath
 
 
@@ -51,10 +52,31 @@ def normalize_archive_parts(member_name: str) -> list[str]:
 
 
 def make_targz(base: str, root_dir: str, base_dir: str) -> str:
-    """Create ``<base>.tar.gz`` of ``root_dir/base_dir`` in GNU tar format."""
+    """Create ``<base>.tar.gz`` of ``root_dir/base_dir`` in GNU tar format.
+
+    Writes to a sibling temp file and renames onto ``archive_path`` only
+    after the archive is fully written. ``tarfile.open`` on a path truncates
+    the destination the instant it opens, so writing there directly means a
+    failure partway through ``tf.add`` (disk full, permission loss,
+    interruption) destroys whatever was already at that path — including an
+    existing export the caller chose to overwrite.
+    """
     archive_path = f"{base}.tar.gz"
-    with tarfile.open(archive_path, "w:gz", format=tarfile.GNU_FORMAT) as tf:
-        tf.add(str(Path(root_dir) / base_dir), arcname=base_dir)
+    dest_dir = os.path.dirname(archive_path) or "."
+    fd, tmp_path = tempfile.mkstemp(
+        dir=dest_dir, prefix=".archive_", suffix=".tar.gz.tmp"
+    )
+    try:
+        with os.fdopen(fd, "wb") as f:
+            with tarfile.open(fileobj=f, mode="w:gz", format=tarfile.GNU_FORMAT) as tf:
+                tf.add(str(Path(root_dir) / base_dir), arcname=base_dir)
+        os.replace(tmp_path, archive_path)
+    except BaseException:
+        try:
+            os.unlink(tmp_path)
+        except OSError:
+            pass
+        raise
     return archive_path
 
 
--- a/tests/hermes_cli/test_archive_safe.py
+++ b/tests/hermes_cli/test_archive_safe.py
@@ -1,0 +1,77 @@
+"""Tests for the shared tar.gz writer (``hermes_cli.archive_safe.make_targz``).
+
+``make_targz`` backs both ``hermes profile export`` and ``hermes kanban
+export``. The contract pinned down here: a failure partway through writing
+the archive (disk full, permission loss, interruption) must never destroy a
+pre-existing file at the destination path — the same failure-atomicity
+guarantee the desktop gateway-download path already has.
+"""
+
+from __future__ import annotations
+
+import sys
+import tarfile
+from pathlib import Path
+
+import pytest
+
+_WORKTREE = Path(__file__).resolve().parents[2]
+if str(_WORKTREE) not in sys.path:
+    sys.path.insert(0, str(_WORKTREE))
+
+from hermes_cli.archive_safe import make_targz
+
+
+def _stage_source(tmp_path: Path) -> None:
+    payload = tmp_path / "src" / "inner"
+    payload.mkdir(parents=True)
+    (payload / "file.txt").write_text("hello")
+
+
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed `hermes_cli/archive_safe.py` (+ new tests) for PR #97835. Converting `make_targz` to write a sibling temp file and `os.replace` only on success is the correct failure-atomicity fix: `tarfile.open(path, "w:gz")` truncates the destination on open, so a mid-`add` failure previously destroyed any pre-existing archive. The temp file is placed in the destination directory (same filesystem, so `os.replace` is atomic) and is cleaned up on failure — including the `BaseException` catch which covers interruption. Tests pin all three contracts: failure preserves the sentinel, success round-trips content, and success overwrites.

Non-blocking observations:

1. **`os.replace` atomicity relies on the temp file being on the same filesystem as `archive_path`** — it is (same directory via `mkstemp(dir=dest_dir)`), so this is solid. If `base` ever points across volumes, `os.replace` would fail and `make_targz` would raise after having written a valid temp file that gets unlinked in the `except` — acceptable, but the unlink-on-`os.replace`-failure path is worth a glance.

2. **Minor:** the temp file name carries `.tar.gz.tmp` (plus a random suffix); if a crash happens between `mkstemp` and the `try` body, an orphaned `.archive_*` file could remain (the `try/except BaseException` only covers the write+rename block). The test asserts no leftovers *after a handled failure*, which is the contract that matters; an abrupt process kill is inherently un-cleanable in-process.

3. **Minor:** `make_targz`'s caller-facing behavior is unchanged (same return value and same final artifact), so the two consumers (`profile export`, `kanban export`) need no updates — good.

Clean, well-scoped change with a real safety improvement. Nothing blocking.

egilewski:
looks mergeable

The replayed change makes archive creation failure-atomic by writing a same-directory temporary tarball and replacing the destination only after close. Current main loses a pre-existing archive when tar creation fails; replay probes and export tests preserve it, clean temporary files, retain overwrite and round-trip behavior, and expose no alternate writer.

Security evidence:
- trust boundary: Potentially untrusted profile or board contents and export failures enter make_targz through the profile and kanban export callers. The sink is the caller-selected archive path; the required boundary is that a failed export cannot alter an existing destination.
- source/sink/invariant: For an exception while opening, adding, closing, or renaming, prior destination bytes remain unchanged and the temporary sibling is removed. On success the destination is a valid GNU tar.gz containing the staged tree.
- current-main reproduction: A bounded probe against current main forced TarFile.add to fail after the destination opened: the exception propagated but the sentinel was replaced by a 28-byte partial tar stream. A normal success round-trip still worked.
- PR-head or patch-replay validation: The controller replay index exactly matches the reviewed PR head for both changed files. Replay tests passed 3/3; a probe covering post-write add failure and os.replace failure reported preserved sentinel bytes, no temporary leftovers, and a successful round-trip.
- positive/negative cases: Positive cases covered archive round-trip and overwrite, two profile-export tests, four credential-export tests, and 19 kanban-transfer tests. Negative cases forced add and rename failures; both preserved prior bytes and cleaned temporary files.
- residual bypass search: All repository call sites route through make_targz for profile and kanban export; no duplicate writer was found. mkstemp creates the temporary file in the destination directory, and os.replace runs only after tar close. Abrupt process loss can leave a private temporary file but cannot replace the destination.
- reviewer validation: CodeRabbit completed with no findings in the clean-pass flow.

teknium1:
Verified against `origin/main` @ `26f178e5fa`: `hermes_cli/archive_safe.py:make_targz` still opens the destination directly with `tarfile.open(archive_path, "w:gz")`, so the bug is real and this is not redundant. The mkstemp-in-dest-dir → `os.replace` pattern guarantees same-filesystem atomic rename, matches the convention in `agent/secret_sources/_cache.py:193-205`, and the `BaseException` cleanup handles KeyboardInterrupt too. Both call sites (`hermes_cli/profiles.py:2260,2275` and `hermes_cli/kanban_transfer.py:229`) go through this function and neither does its own temp-staging of the archive file, so the fix covers the whole class. Test quality is good (mid-write failure preserves sentinel + asserts no leftover temp file).

One behavioral change the diff doesn't mention: `tempfile.mkstemp` creates the file `0600`, and `os.replace` preserves that mode — so exported archives change from umask-default (typically `0644`) to `0600`. Neither call site chmods the result afterward (no `chmod` after `make_targz` in `profiles.py` or `kanban_transfer.py`). For profile exports that's arguably an improvement; for kanban board exports intended to be handed to someone else it's a silent tightening that could surprise scripted pipelines reading the archive as another user. Worth either an explicit `os.chmod(tmp_path, 0o644 & ~umask)`-style normalization before the replace, or a deliberate note in the docstring that exports are now owner-only.

Minor: no `fsync` before the replace, but the existing codebase pattern (`_cache.py`, `cron/jobs.py`) doesn't fsync either, so that's consistent with local convention.

Verdict: correct and merge-ready once the file-mode change is either intentional-and-documented or normalized.

alt-glitch:

Superseded by the active salvage #98018, which carries this same failure-atomic make_targz repair. The data-integrity issue remains important; this duplicate link is only to keep review on the current implementation.

PR #77547, opened 2026-08-03
fix(slack): preserve long clarify option text

hrsong99:
## Summary
- render clarify options wider than Slack clients can display as full numbered sections
- count Korean/CJK characters by display width and switch layout above 48 columns
- keep controls compact with `Select N` accessory buttons
- split choices safely across Slack’s 3,000-character section limit without breaking escaped entities
- show durable `Choice received; request resumed` feedback after selection
- sanitize callback updates so Slack removes inert buttons reliably

## Reproduction
Ouroboros/PRD interviews can generate multi-sentence Korean options. Slack accepts button labels up to 75 characters but visually ellipsizes them much earlier, hiding decision-critical text. The screenshot reproduced this with 29–50-character Korean choices measuring 51–82 display columns.

## Verification
- `uv run --extra dev --extra slack pytest tests/gateway/test_slack.py -o addopts= -q` (**171 passed**)
- exact screenshot choices trigger the full-text layout
- regression coverage for CJK under 75 characters, >3,000-character choices, escaped entities, callback updates, and resumed-state feedback
- `uv run --extra dev --extra slack ruff check plugins/platforms/slack/adapter.py tests/gateway/test_slack.py`
- `git diff --check origin/main...HEAD`

Rebased onto current `origin/main` on 2026-08-21 while preserving the original commits/authorship.

PR #97836, opened 2026-08-29
fix(slack): sanitize resolved clarify message blocks

### LINKED ISSUE #97823: [Bug]: Slack clarify buttons remain active after long-prompt choice resolves
## Bug Description

Slack `clarify` buttons successfully resolve the user's choice, but the message can fail to update afterward when the original question exceeds Slack's 3,000-character section-text limit. The agent resumes and receives the answer, while the buttons remain visible and appear unresponsive.

This is the clarify-path equivalent of the approval-card bug previously reported in #53693 and #62054. Those issues were fixed for approval cards by routing `chat.update` blocks through `sanitize_blocks()`, but the sibling `_update_clarify_message()` path still sends `updated_blocks` unsanitized.

## Steps to Reproduce

1. Run the Hermes gateway with Slack enabled.
2. Have the agent call `clarify` with a multiple-choice question whose question text is longer than 3,000 characters.
3. Click one of the Slack Block Kit choice buttons.
4. Observe that the agent receives the selected choice and continues.
5. Observe that the Slack message does not change to the resolved state and its buttons remain visible.

## Expected Behavior

The choice should resolve, the message should update in place with the selected result, and the buttons should be removed.

## Actual Behavior

The choice resolves successfully, but Slack rejects the subsequent `chat.update` call:

```text
[Slack] Failed to update clarify message: The request to the Slack API failed.
The server responded with: {'ok': False, 'error': 'invalid_blocks',
'1': 'must be less than 3001 characters [json-pointer:/blocks/0/text/text]'}
Slack button resolved clarify (...)
```

Because resolution happens before the failed UI update, the agent proceeds while Slack continues displaying apparently active buttons.

## Affected Component

Gateway (Telegram/Discord/Slack/WhatsApp)

Slack

## Debug Report

```shell
Not uploaded. The exact relevant gateway error and live reproduction details are included above; no wider diagnostic bundle is needed to reproduce the deterministic Block Kit boundary failure.
```

## Operating System

macOS 26.5.2

## Python Version

3.11.15

## Hermes Version

0.20.6 (2026.8.27), reproduced with a few unrelated local commits on top of upstream `25fcc8ad`; the same unsanitized update path remains on current upstream `main`.

```shell
2026-08-29 11:01:24 WARNING [Slack] Failed to update clarify message: ... invalid_blocks ... must be less than 3001 characters [json-pointer:/blocks/0/text/text]
2026-08-29 11:01:24 INFO Slack button resolved clarify (... choice_index=0 ...)
```

## Root Cause Analysis (optional)

`plugins/platforms/slack/adapter.py::_update_clarify_message()` rebuilds the resolved card using the interaction payload's `question_text`, then calls `chat_update(..., blocks=updated_blocks)` without passing those blocks through `sanitize_blocks()`.

Slack's shared `sanitize_blocks()` boundary already clamps section and context text to 3,000 characters. Approval-card updates use it, which fixed #53693/#62054, but clarify-card updates do not.

## Proposed Fix (optional)

Open PR #77547 may fix this by changing `_update_clarify_message()` to send `sanitize_blocks(updated_blocks)`. It also improves long-choice rendering, but it is based on an older `main` commit and has remained open without checks or review:

- #77547

A focused current-main regression should cover a question body over 3,000 characters—not only a long selected choice—and verify that the callback update removes the buttons with every emitted Block Kit text object within Slack's limits.

will-lynas:
## What does this PR do?

Fixes Slack clarification cards that remain visibly interactive after a choice has already resolved. Slack may return the original question with additional HTML escaping in the interaction payload; reusing that text directly in `chat.update` can exceed Slack's 3,000-character Block Kit limit and cause `invalid_blocks`.

This applies the existing shared `sanitize_blocks()` boundary to resolved clarification-card updates. It does not change how choices are rendered.

## Related Issue

Fixes #97823

Related: #77547

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- Route resolved Slack clarification blocks through `sanitize_blocks()` before `chat.update`.
- Add a producer-shaped regression that starts from a valid sent question, simulates Slack re-escaping it in the interaction payload, and verifies the resolved card removes its actions while respecting the 3,000-character limit.

## How to Test

1. Run `uv run --frozen --extra dev --extra slack pytest tests/gateway/test_slack.py tests/gateway/test_slack_clarify_buttons.py -q`.
2. Run `uv run --frozen --extra dev ruff check plugins/platforms/slack/adapter.py tests/gateway/test_slack_clarify_buttons.py`.
3. Run `uv run --frozen python scripts/check-windows-footguns.py --all`.

The new regression fails on the previous implementation with a 5,376-character update section and passes after the sanitizer is applied.

### Code

- [x] I've tested on my platform: macOS 26.5.2

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A: no documentation change required
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — or N/A: no config keys changed
- [x] I've updated `CONTRIBUTING.md` or `AGENTS.md` if I changed architecture or workflows — or N/A: no architecture or workflow changes
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#compatibility-guide) — Windows footgun scan passed
- [x] I've updated tool descriptions/schemas if I changed tool behavior — or N/A: no model-facing tool behavior changed

## Screenshots / Logs

Focused affected suites: **240 passed**. Ruff, `git diff --check`, and the Windows-footgun scan also passed.

### DIFF
--- a/plugins/platforms/slack/adapter.py
+++ b/plugins/platforms/slack/adapter.py
@@ -7861,7 +7861,7 @@ class SlackAdapter(BasePlatformAdapter):
                 channel=channel_id,
                 ts=msg_ts,
                 text=decision_text,
-                blocks=updated_blocks,
+                blocks=sanitize_blocks(updated_blocks),
             )
         except Exception as e:
             logger.warning("[Slack] Failed to update clarify message: %s", e)
--- a/tests/gateway/test_slack_clarify_buttons.py
+++ b/tests/gateway/test_slack_clarify_buttons.py
@@ -186,6 +186,63 @@ class TestSlackClarifyChoiceAction:
         assert entry is not None
         assert not entry.event.is_set()
 
+    @pytest.mark.asyncio
+    async def test_choice_update_sanitizes_slack_reescaped_question(self):
+        from plugins.platforms.slack.block_kit import MAX_SECTION_TEXT
+        from tools import clarify_gateway as cm
+
+        adapter = _make_adapter()
+        _attach_auth_runner(adapter)
+        mock_client = adapter._team_clients["T1"]
+        mock_client.chat_postMessage = AsyncMock(return_value={"ts": "3.3"})
+        mock_client.chat_update = AsyncMock(return_value={"ok": True})
+
+        question = "Approve this exact content? " + ("&" * 700)
+        await adapter.send_clarify(
+            chat_id="C1",
+            question=question,
+            choices=["Approve", "Reject"],
+            clarify_id="cid-reescaped-question",
+            session_key="sk-reescaped-question",
+        )
+        cm.register(
+            "cid-reescaped-question",
+            "sk-reescaped-question",
+            question,
+            ["Approve", "Reject"],
+        )
+
+        sent_blocks = mock_client.chat_postMessage.await_args.kwargs["blocks"]
+        sent_question = sent_blocks[0]["text"]["text"]
+        assert len(sent_question) == MAX_SECTION_TEXT
+        interaction_question = sent_question.replace("&", "&amp;")
+        assert len(interaction_question) > MAX_SECTION_TEXT
+
+        body = {
+            "message": {
+                "ts": "3.3",
+                "blocks": [
+                    {
+                        "type": "section",
+                        "text": {"type": "mrkdwn", "text": interaction_question},
+                    },
+                    {"type": "actions", "elements": [{"type": "button"}]},
+                ],
+            },
+            "channel": {"id": "C1"},
+            "user": {"name": "will", "id": "U_WILL"},
+        }
+        action = {
+            "action_id": "hermes_clarify_choice_0",
+            "value": "cid-reescaped-question|0",
+        }
+
+        await adapter._handle_clarify_action(AsyncMock(), body, action)
+
+        updated_blocks = mock_client.chat_update.await_args.kwargs["blocks"]
+        assert all(block["type"] != "actions" for block in updated_blocks)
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed `plugins/platforms/slack/adapter.py` (+ test) for PR #97836. Applying the existing `sanitize_blocks()` boundary to resolved clarification-card updates is the minimal, correct fix: Slack's interaction payload can re-escape the question text (e.g. `&` → `&amp;`), pushing the blocks past the 3,000-char / `MAX_SECTION_TEXT` limit and causing `invalid_blocks` on `chat.update`. The regression test is well constructed — it drives the re-escaped question past `MAX_SECTION_TEXT`, verifies the actions block is removed on resolution, and asserts the updated section text is back within the limit.

Non-blocking observation:

- **Verify `sanitize_blocks` is already imported in `adapter.py`.** The diff only adds the call site, not the import. If the module-level import already exists (it's described as an existing shared helper used elsewhere in the file), no action needed; if not, the module won't compile. Worth a quick confirm.
- **Minor:** the fix sanitizes `updated_blocks` but leaves `decision_text` (the plain-text `text=` argument to `chat_update`) unsanitized. If the re-escaped text is also passed there it's not subject to the Block Kit limit, so this is fine — just confirming the text path was intentionally left as-is.

Small, well-targeted fix with a precise regression test. Nothing blocking.

PR #97837, opened 2026-08-29
fix(model-metadata): resolve routed OpenRouter context windows correctly

### LINKED ISSUE #97820: [Bug]: OpenRouter routing suffixes (:floor/:nitro/:free) break models.dev catalog lookup — context window falls back to hardcoded family default
## Summary

When an OpenRouter model ID carries a routing-variant suffix (`:floor`, `:nitro`, `:free`, `:exacto`, …), Hermes fails the models.dev catalog lookup for that model and falls through to the hardcoded family-default table in `agent/model_metadata.py`. The resolved context window is then silently wrong — in my case `z-ai/glm-5.3-flash:floor` resolves to **202,752** instead of the catalog value **1,310,720**.

Routing suffixes only change how OpenRouter picks a provider (price/throughput sorting); they do not change the model. The catalog lookup should strip them before matching.

## Environment

- Hermes Agent v0.20.1 (2026.8.13), commit `56a41715dc` (git install)
- Python 3.11.15, Linux (WSL2)
- Provider: `openrouter`, `model.default: z-ai/glm-5.3-flash:floor`

## Steps to reproduce

1. Set `model.default: z-ai/glm-5.3-flash:floor` in `config.yaml`.
2. Start a session; the UI shows the context window as **202.8k**.
3. The same model without the suffix shows the correct value.

### Minimal repro (verified locally)

```python
from agent.models_dev import _find_model_entry, _get_provider_models

m1 = get_model_context_length('z-ai/glm-5.3-flash:floor', provider='openrouter')
m2 = get_model_context_length('z-ai/glm-5.3-flash', provider='openrouter')
print('suffixed :floor ->', m1)   # 202752   (wrong)
print('bare            ->', m2)   # 1310720  (correct, matches models.dev)

models = _get_provider_models('openrouter', allow_network=False)
print(_find_model_entry(models, 'z-ai/glm-5.3-flash:floor'))  # None        (miss)
print(_find_model_entry(models, 'z-ai/glm-5.3-flash'))        # entry found, limit.context == 1310720
```

Output on my machine:

```
suffixed  :floor -> 202752
bare            -> 1310720
catalog entry (suffixed): MISS
catalog entry (bare)    : FOUND
bare entry limit: {'context': 1310720, 'output': 131072}
```

## Root cause

1. `agent/models_dev.py` → `_find_model_entry()` (~line 1130): the suffix-aware fallback only tries `:cloud` / `-cloud` (line 1153). OpenRouter routing suffixes are not stripped, so a suffixed ID misses the catalog even when the bare slug is present.
2. `agent/model_metadata.py` → `get_model_context_length()` (~line 2568): resolution order is config override → persistent probe cache → custom providers → … → models.dev lookup (step 5f, same `:cloud`/`-cloud`-only fallback) → OpenRouter live API (Kimi guard only) → hardcoded family defaults (step 8). A suffixed model falls all the way to step 8.
3. Step 8's family table (lines ~484–491) has `"glm-5.2": 1_048_576` but the generic fallback `"glm": 202752`. `glm-5.3-flash` matches `"glm"` → 202752.
4. WebUI's `_fmtTokens` renders 202752 as `202.8k`, which is how I noticed.

models.dev and OpenRouter both list the real value: models.dev entry `z-ai/glm-5.3-flash` has `limit.context = 1310720`; OpenRouter's endpoints API shows 1,048,576 on most providers (Cloudflare 1,310,720).

## Impact

- Not cosmetic: this value feeds context-pressure / compaction thresholds, so sessions compact far earlier than they should (202,752 assumed vs ≥1,048,576 actually available on every cheap endpoint that `:floor` routes to).
- Generalizes to any OpenRouter-routed model ID with a routing suffix whose true spec differs from the hardcoded family default. E.g. a hypothetical `glm-5.2:floor` would work by luck, while `glm-5.3-anything:suffix` gets 202752.

## Workaround

`model.context_length` config override wins at resolution step 0, so this unblocks per-machine:

```yaml
model:
  context_length: 1048576
```

(`hermes config set model.context_length 1048576` — I used the max context actually offered by the `:floor`-routed endpoints rather than the 1.25M Cloudflare-only value.)

## Suggested fix

Normalize the model ID before the catalog lookup: strip a trailing OpenRouter routing-variant suffix (`:free`, `:extended`, `:thinking`, `:nitro`, `:floor`, `:exacto`, `:online`) in `_find_model_entry()` (or in the step-5f caller in `get_model_context_length()`), keeping the full suffixed ID for provider-facing config keys. That keeps the existing `:cloud`/`-cloud` behavior intact and makes every catalog consumer (`get_model_capabilities`, `lookup_models_dev_context`) agree on what "this model is in the catalog" means.

fangliquanflq:
## What does this PR do?

OpenRouter users who select a routed model ID such as `z-ai/glm-5.3-flash:floor` now receive the underlying model's models.dev metadata instead of a hardcoded family fallback. This keeps context pressure and compaction thresholds aligned with the catalog while preserving the full routed ID for provider requests and model info.

### Symptom

`get_model_context_length("z-ai/glm-5.3-flash:floor", provider="openrouter")` resolved to 202,752 tokens, while the bare catalog model resolved to 1,310,720 tokens.

### Impact

Affected OpenRouter sessions can compact much earlier than the selected model's catalog context window requires. Capability and rich model-info lookups also treat routed IDs as unknown even when the underlying model exists in models.dev.

### Bug Cause

**Trigger:** `agent/models_dev.py` / `lookup_models_dev_context()` and `_find_model_entry()` when an OpenRouter model ID ends in a routing suffix.

**Causal chain:**
1. OpenRouter passes a full routed model ID such as `z-ai/glm-5.3-flash:floor` into metadata resolution.
2. Catalog matching only checks the full ID plus the Ollama-specific `:cloud` and `-cloud` variants, so the bare models.dev entry is missed.
3. Context resolution falls through to the generic GLM family default and reports a much smaller window.

**Why it is wrong:** OpenRouter routing suffixes select provider-routing behavior; they do not identify a different underlying model for catalog metadata.

**Working sibling / contrast:** The same model without the routing suffix matches models.dev and resolves its catalog metadata correctly.

**Ruled out:** This is not missing catalog data. The bare `z-ai/glm-5.3-flash` entry is present and the regression test proves that only the routed identity misses before normalization.

### Fix

Catalog lookup now strips only the documented OpenRouter routing suffixes and only for the OpenRouter provider. Context, capability, and rich model-info paths share this provider-aware lookup, while unknown suffixes and non-OpenRouter providers retain exact matching behavior. The provider-facing model ID remains unchanged.

## Related Issue

Closes #97820

## Type of Change

## Changes Made

- `agent/models_dev.py` - centralize provider-aware catalog matching and normalize OpenRouter routing suffixes for metadata lookup.
- `tests/agent/test_models_dev.py` - cover all supported routing suffixes, provider scoping, unknown suffixes, capabilities, and full-ID preservation.
- `tests/agent/test_model_metadata.py` - exercise the end-to-end context-length resolution path for a routed OpenRouter model.

## How to Test

1. Run the related metadata suites:

```bash
scripts/run_tests.sh tests/agent/test_models_dev.py tests/agent/test_model_metadata.py -q
```

2. Confirm all 190 tests pass, including the routed `z-ai/glm-5.3-flash:floor` context regression.

### Code

- [x] I've run the related test suites and all tests pass

- [x] Relevant documentation update is N/A; behavior and supported suffixes are documented in code and regression tests
- [x] `CONTRIBUTING.md` and `AGENTS.md` updates are N/A; no architecture or workflow changed
- [x] Cross-platform impact was considered; matching is pure string logic with platform-independent tests
- [x] Tool description/schema updates are N/A; no tool behavior changed

### DIFF
--- a/agent/models_dev.py
+++ b/agent/models_dev.py
@@ -774,42 +774,11 @@ def lookup_models_dev_context(
     if not isinstance(models, dict):
         return _default_override_context(provider)
 
-    # Exact match
-    entry = models.get(model)
-    if entry:
+    entry = _find_model_entry(models, model, provider=provider)
+    if entry is not None:
         ctx = _extract_context(entry)
         if ctx:
             return ctx
-
-    # Case-insensitive match
-    model_lower = model.lower()
-    for mid, mdata in models.items():
-        if mid.lower() == model_lower:
-            ctx = _extract_context(mdata)
-            if ctx:
-                return ctx
-
-    # Suffix-aware fallback: some providers (e.g. ollama-cloud) store
-    # model IDs with :cloud / -cloud suffixes in models.dev while the
-    # live API returns bare names.  Without this, kimi-k2.6 misses the
-    # kimi-k2.6:cloud entry and falls through to stale OpenRouter metadata
-    # reporting 32768 — tripping the 64k minimum-context guard.
-    # The suffix-stripping in fetch_ollama_cloud_models() handles the
-    # model-picker UX; this handles the context-length lookup path.
-    for suffix in (":cloud", "-cloud"):
-        suffixed_key = model + suffix
-        entry = models.get(suffixed_key)
-        if entry:
-            ctx = _extract_context(entry)
-            if ctx:
-                return ctx
-        # Also try case-insensitive
-        suffixed_lower = model_lower + suffix
-        for mid, mdata in models.items():
-            if mid.lower() == suffixed_lower:
-                ctx = _extract_context(mdata)
-                if ctx:
-                    return ctx
 
     # Catalog miss — a _default override may fill the gap (#84482).
     return _default_override_context(provider)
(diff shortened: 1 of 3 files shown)


Enough1122:

Reviewed `agent/models_dev.py` (+ tests) for PR #97837. The refactor from the scattered exact/case-insensitive/suffix fallbacks into `_catalog_model_id` + `_find_model_entry_with_id` is a genuine improvement: provider-aware stripping of OpenRouter routing suffixes before the catalog lookup, while preserving the full routed ID for the provider/config layer (`get_model_info` returns the routed ID). The `:cloud`/`-cloud` fallback is preserved (folded into the shared helper) so the kimi-k2.6:cloud path still works, and tests cover every documented suffix plus the provider-scoped and unknown-suffix negatives.

Non-blocking observations:

1. **This overlaps heavily with #97839, which fixes the same symptom in the same files** (`agent/models_dev.py` `_find_model_entry` / `lookup_models_dev_context` / `get_model_capabilities` / `get_model_info`, plus `agent/model_metadata.py`). #97839 also strips in `model_metadata.py`'s `get_model_context_length` directly, which #97837 does not touch (it relies on `lookup_models_dev_context` for that path). Whichever lands first will conflict with the other; the maintainers should pick one and reconcile, ideally sharing one definition of the routing-suffix set and one strip helper rather than the duplicated constants both PRs introduce.

2. **Case handling in `_catalog_model_id` is exact-lowercase for the provider id** — fine. For the suffix strip it uses `model.lower().endswith(suffix)` which is case-insensitive; note `model[: -len(suffix)]` then returns the original-case prefix (correct — preserves casing of the real model id). Good.

3. **Minor:** `_find_model_entry_with_id` returns a tuple while `_find_model_entry` unwraps it — reasonable, but the two-function split adds surface; the tests exercise both `lookup_models_dev_context` and `get_model_capabilities`/`get_model_info` through it, so coverage is adequate.

4. **Minor:** `_default_override_context(provider)` still runs when all candidate lookups miss — the routed model falling through to a hardcoded family default is now only possible when the bare entry is also absent from the catalog, which is the correct residual behavior.

Solid refactor with good tests. The only real action item is coordinating with #97839.

fangliquanflq:
Thanks for the detailed review. I rechecked each observation against the current head and `main`:

1. The overlap with #97839 is confirmed. #97837 remains intentionally scoped to #97820’s models.dev catalog miss. `get_model_context_length()` already falls through from an exact OpenRouter live-metadata lookup to `lookup_models_dev_context()`, where this PR performs the provider-scoped normalization. Adding the same suffix table to `model_metadata.py` would duplicate the definition without being needed for the reported catalog-backed case. Maintainers can therefore choose between the overlapping implementations; no code change is needed in this PR for coordination.
2. Confirmed: suffix matching is case-insensitive while the original model prefix is preserved.
3. Confirmed: the tuple-returning helper is used only where the matched catalog ID is required to preserve the routed provider-facing ID; other callers use the entry-only wrapper.
4. Confirmed: `_default_override_context()` remains a catalog-miss fallback and cannot preempt a routed model whose bare catalog entry exists.

Evidence: `tests/agent/test_models_dev.py` covers all documented suffixes, provider scoping, unknown suffixes, capabilities, and routed-ID preservation; `tests/agent/test_model_metadata.py` exercises the full context-resolution path through the models.dev fallback. At head `3fe122df177549ed3f67eb552187a2f2e7f07ffb`, both “Python tests / Run tests” and “All required checks pass” are successful.

teknium1:
Closing: #97820 is fixed on main via #108666 (1021a0325696, salvaged from #100651 by @jakobdylanc). The chosen implementation strips only `:nitro`/`:floor`/`:exacto`/`:online`; `:free`/`:batch` are real catalog SKUs with their own context windows (13 differ from their base, e.g. `z-ai/glm-5.2:free` 256K vs 1.05M), so stripping them would over-report the window. Thanks for the report and the fix here.

Issue #94138, opened 2026-08-24
redact_sensitive_text corrupts binary/base64 tool output (CDP screenshots), surfacing as provider 400 invalid_image


leduncfresh:
`_redact_cdp_output` (tools/browser_cdp_tool.py:527) applies `redact_sensitive_text(force=True)` to `Page.captureScreenshot` results. The `gAAAA[A-Za-z0-9_=-]{20,}` pattern matches inside PNG base64 wherever `gAAAA` follows `+`/`/`, collapsing arbitrary-length spans to `first6...last4` (short matches → `***`). Files decoded from that output are corrupt; the persisted full copies in tool_result_storage are redacted too, so data is unrecoverable. Downstream, `vision_analyze` embeds the corrupt PNGs and the provider rejects with 400 `invalid_image`, which the conversation loop reports as "model provider failed after retries" — misleading operators toward provider/auth debugging.

Repro: `redact_sensitive_text("gAAAA" + "B"*60, force=True)` → `gAAAAB...BBBB`.

Suggested fix: skip redaction for data-URI/base64-image fields (or all high-volume binary payloads), and/or validate images in the vision path before embedding with a clear per-file error.

**Expected:** screenshot base64 passes through tool output byte-identical; secret redaction does not fire inside binary/base64 image payloads.
**Actual:** base64 spans matching `gAAAA…` are collapsed in-place; decoded PNGs have valid headers but CRC failures mid-IDAT and no IEND. Live impact in our deployment: 6 of 14 captured screenshots corrupted, two resume sessions killed by the resulting deterministic 400.

**Environment:** Hermes Agent v0.20.0 (2026.8.3, git install, unmodified), macOS, model `grok-4.5`, provider `xai-oauth`.

**Related (checked, believed not duplicates):** #5322 (redaction corrupting source files on disk — different path), #363 (file-tool redaction output gap), #69078 (same xAI 400 `invalid_image` symptom but with a verified-valid source image — that one is about session-bricking recovery; here the image really is corrupt, and the redactor is what corrupted it).

PR #97838, opened 2026-08-29
fix(browser): add getResponseBodyForInterception and getResourceContent to the CDP binary-redaction table

nftpoetrist:
## Summary

`#97078` (merged) built `_CDP_ALWAYS_BINARY_PATHS`/`_CDP_FLAGGED_BINARY_PATHS` in `tools/browser_cdp_tool.py` so `redact_sensitive_text`'s Fernet-shaped pattern (`gAAAA[A-Za-z0-9_=-]{20,}`) can't corrupt opaque base64 payloads returned by CDP calls (#94138). Four methods were covered: `Network.getResponseBody`, `Fetch.getResponseBody`, `IO.read`, `Network.getRequestPostData`.

Two protocol methods with the exact same `{field, base64Encoded}` shape were missed:
- `Network.getResponseBodyForInterception` → `{body, base64Encoded}`
- `Page.getResourceContent` → `{content, base64Encoded}`

Both are deprecated-but-still-live CDP methods a model can reach through the raw `browser_cdp` passthrough tool. A binary payload returned through either currently falls through to full-text redaction and gets corrupted the same way the four already-covered methods did before #97078.

## Fix

Add both methods to `_CDP_FLAGGED_BINARY_PATHS` with their respective carrier field, following the exact same pattern as the four existing entries.

## Test plan

- [x] Added 4 new tests mirroring the existing `Network.getResponseBody`/`Fetch.getResponseBody` coverage: a base64-flagged pass-through test and a text-discriminator-still-redacts test for each of the two methods.
- [x] Mutation-verified: stashed the fix, confirmed the two pass-through tests fail against the original code — the base64 payload gets corrupted from `/gAAAAEEEE...EEEE==` down to a truncated `/gAAAAE...EE==` mask, reproducing the exact corruption #94138 described. Restored the fix, confirmed both pass again.
  - Note: the redaction pattern requires a non-alphanumeric character immediately before `gAAAA` (regex lookbehind), so the fixture uses a `/` prefix rather than an arbitrary alphanumeric one — an earlier fixture attempt with an alphanumeric prefix didn't actually exercise the redaction path and would have passed regardless of the fix.
- [x] Full `tests/tools/test_browser_cdp_tool.py` + `tests/tools/test_browser_cdp_override.py` (44 tests) pass unchanged.

### DIFF
--- a/tests/tools/test_browser_cdp_tool.py
+++ b/tests/tools/test_browser_cdp_tool.py
@@ -338,6 +338,75 @@ def test_fetch_get_response_body_base64_discriminator_passes_through(cdp_server)
     assert result["result"]["body"] == body_b64
 
 
+def test_get_response_body_for_interception_base64_discriminator_passes_through(cdp_server):
+    """Network.getResponseBodyForInterception pins the same body/base64Encoded
+    contract as Network.getResponseBody — same protocol-defined shape,
+    deprecated but still a live method (#94138 sibling gap)."""
+    body_b64 = "/" + "gAAAA" + "E" * 60 + "=="
+    cdp_server.on(
+        "Network.getResponseBodyForInterception",
+        lambda params, sid: {"body": body_b64, "base64Encoded": True},
+    )
+
+    result = json.loads(
+        browser_cdp_tool.browser_cdp(method="Network.getResponseBodyForInterception")
+    )
+
+    assert result["success"] is True
+    assert result["result"]["body"] == body_b64
+
+
+def test_get_response_body_for_interception_text_discriminator_still_redacts(cdp_server):
+    """Same method with base64Encoded: false — the body is text and a real
+    secret in it must still be redacted."""
+    fake_key = "sk-" + "CDPINTERCEPTBODY1234567890"
+    cdp_server.on(
+        "Network.getResponseBodyForInterception",
+        lambda params, sid: {"body": f"leak {fake_key} here", "base64Encoded": False},
+    )
+
+    result = json.loads(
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed `tools/browser_cdp_tool.py` (+ tests) for PR #97838. This is a correct, minimal extension of the #97078 binary-redaction table: `Network.getResponseBodyForInterception` and `Page.getResourceContent` carry the same `{field, base64Encoded}` shape as the already-covered methods, so adding them to `_CDP_ALWAYS_BINARY_PATHS` closes a real redaction gap. The four tests pin both sides of the contract — base64 payloads pass through untouched (Fernet-shaped pattern can't corrupt them) while `base64Encoded: false` text still gets redacted.

Non-blocking observations:

1. **The `_CDP_FLAGGED_BINARY_PATHS` counterpart was not extended.** The diff only adds to `_CDP_ALWAYS_BINARY_PATHS`. If `Page.getResourceContent`/`Network.getResponseBodyForInterception` should also be in the "flagged" (conditional) table for some consumers, that's a separate gap; the PR description only claims the always-binary table, so this is likely intentional — worth a one-line note for completeness.

2. **Minor:** `Page.getResourceContent`'s carrier key is `content` while the others use `body` — the table correctly encodes per-method field names, and the tests confirm the right key is read. No issue.

3. **Minor:** the protocol methods added are documented/deprecated in places (`getResponseBodyForInterception` is legacy but still live per the test comment) — fine to cover; no action needed.

Small, well-tested, low-risk addition that follows the established pattern exactly. Nothing blocking.

PR #97839, opened 2026-08-29
fix(openrouter): strip routing suffixes before models.dev catalog lookup

### LINKED ISSUE #97820: [Bug]: OpenRouter routing suffixes (:floor/:nitro/:free) break models.dev catalog lookup — context window falls back to hardcoded family default

Finn763:
Closes #97820

## What Problem This Solves

OpenRouter routing suffixes (`:free`, `:extended`, `:thinking`, `:nitro`, `:floor`, `:exacto`, `:online`) only change provider routing, not the model identity. Hermes looked up `z-ai/glm-5.3-flash:floor` verbatim in `models.dev` and `fetch_model_metadata()` and fell through to the hardcoded family default `glm: 202752`, so the UI showed 202.8k instead of the catalog value **1,310,720**. This feeds compaction thresholds, so sessions compact far too early.

`agent/models_dev.py:_find_model_entry()` only handled `:cloud`/`-cloud`, and `agent/model_metadata.py:get_model_context_length()` had the same blind spot at step 5f. Every consumer (`lookup_models_dev_context`, `get_model_capabilities`, `get_model_info`, OpenRouter live metadata) missed the suffixed ID.

## Solution

- Introduce whitelist `_OPENROUTER_ROUTING_SUFFIXES` (`free/extended/thinking/nitro/floor/exacto/online`) and helpers `_strip_openrouter_routing_suffix` / `_is_openrouter_provider` / `_should_strip_openrouter_suffix`.
- In `agent/models_dev.py`: `_find_model_entry()`, `lookup_models_dev_context()`, `get_model_info()` now try `[bare, stripped]` candidates (exact + case-insensitive) for OpenRouter only; keeps `:cloud`/`-cloud` fallback. `get_model_capabilities()` propagates `provider`.
- In `agent/model_metadata.py`: same whitelist/helpers, strip suffix before the `fetch_model_metadata()` lookup (both exact and case-insensitive retries) in `get_model_context_length()` step 5f. Full suffixed ID is still used for the wire/config layer.

Non-OpenRouter providers are never stripped.

## Evidence

Minimal repro before fix (`HERMES_HOME` isolated, `fake registry z-ai/glm-5.3-flash=1310720`):

```
_find_model_entry(:floor)=MISS, lookup=None, get_model_capabilities=None
get_model_context_length(:floor)=1048576 (family fallback) vs bare=1310720
```

After fix:

```
_find_model_entry(:floor)=FOUND, lookup=1310720, caps=1310720, get_model_context_length(:floor)=1310720 == bare
```

Boundary `repro_boundary.py` (27 checks) all OK:
- `:free/:nitro/:floor/...` hit, `:unknown` correctly MISS, `:FLOOR` case-insensitive, `anthropic:floor` correctly NOT stripped (None), `kimi-k2.6:floor` finds `:cloud` variant, `base_url` openrouter stripping works.

## Tests

- `pytest tests/agent/test_models_dev.py tests/agent/test_model_metadata.py -k "model_metadata or models_dev"` → **179 passed** (stash before/after identical, no new failures)
- Repro scripts retained untracked as `repro_97820.py` / `repro_boundary.py` for reviewers.

## Risk

Low. Pure lookup normalization, OpenRouter-scoped, no wire-format change. Non-OpenRouter paths unchanged.

## Checklist

- [x] Tested `openrouter:floor` resolves to bare catalog entry, non-openrouter not stripped
- [x] Existing `:cloud` fallback preserved
- [x] No config/wire ID mutated

### DIFF
--- a/agent/model_metadata.py
+++ b/agent/model_metadata.py
@@ -142,6 +142,36 @@ def _strip_provider_prefix(model: str) -> str:
         return suffix
     return model
 
+
+# OpenRouter routing-variant suffixes — same list as agent/models_dev.py.
+# Stripped before catalog lookups so a suffixed ID (``:floor`` etc) hits the
+# bare model entry; the full ID is kept for the provider/config layer.
+_OPENROUTER_ROUTING_SUFFIXES = frozenset({
+    "free", "extended", "thinking", "nitro", "floor", "exacto", "online",
+})
+
+
+def _strip_openrouter_routing_suffix(model: str) -> str:
+    """Strip a trailing OpenRouter routing suffix (``:floor`` etc) if present."""
+    if not isinstance(model, str) or ":" not in model:
+        return model
+    prefix, suffix = model.rsplit(":", 1)
+    if not prefix or not suffix:
+        return model
+    if suffix.strip().lower() in _OPENROUTER_ROUTING_SUFFIXES:
+        return prefix
+    return model
+
+
+def _is_openrouter_provider_name(provider: str) -> bool:
+    """True when *provider* (Hermes id or ``effective_provider``) is OpenRouter."""
+    if not provider or not isinstance(provider, str):
+        return False
+    key = provider.strip().lower()
+    if not key:
+        return False
+    return key == "openrouter"
+
 _model_metadata_cache: Dict[str, Dict[str, Any]] = {}
 _model_metadata_cache_time: float = 0
 _novita_metadata_cache: Dict[str, Dict[str, Any]] = {}
@@ -3384,6 +3414,20 @@ def get_model_context_length(
     if effective_provider == "openrouter":
         metadata = fetch_model_metadata()
         entry = metadata.get(model)
+        # Strip OpenRouter routing suffix (e.g. :floor/:nitro/:free) before
+        # the catalog lookup so a suffixed ID hits the bare entry. The full
+        # ID is kept for the wire/config layer.
+        if not entry:
+            stripped = _strip_openrouter_routing_suffix(model)
+            if stripped != model:
+                entry = metadata.get(stripped)
+                # Also try case-insensitive for the stripped variant
+                if not entry:
+                    stripped_lower = stripped.lower()
+                    for k, v in metadata.items():
+                        if k.lower() == stripped_lower:
+                            entry = v
+                            break
         if entry:
             or_ctx = entry.get("context_length")
             # Guard against the known OpenRouter Kimi-family 32k underreport
@@ -3392,6 +3436,20 @@ def get_model_context_length(
                 or_ctx == 32768 and _model_name_suggests_kimi(model)
             ):
                 return or_ctx
+        # Fallback: stripped model may have different case in metadata keys
+        # (e.g. Z-AI/GLM vs z-ai/glm). Try case-insensitive bare lookup once more.
+        if not entry:
+            model_lower = model.lower()
+            stripped_lower = _strip_openrouter_routing_suffix(model).lower()
+            if stripped_lower != model_lower:
+                for k, v in metadata.items():
+                    if k.lower() == stripped_lower:
+                        or_ctx = v.get("context_length")
+                        if isinstance(or_ctx, int) and or_ctx > 0 and not (
+                            or_ctx == 32768 and _model_name_suggests_kimi(model)
+                        ):
+                            return or_ctx
+                        break
 
     if effective_provider:
         from agent.models_dev import lookup_models_dev_context
--- a/agent/models_dev.py
+++ b/agent/models_dev.py
@@ -214,6 +214,59 @@ PROVIDER_TO_MODELS_DEV: Dict[str, str] = {
     "ollama-cloud": "ollama-cloud",
 }
 
+# OpenRouter routing-variant suffixes that do not change the model identity.
+# These are stripped BEFORE the models.dev catalog lookup so that e.g.
+# ``z-ai/glm-5.3-flash:floor`` resolves to the same entry as the bare slug.
+# Source: https://openrouter.ai/docs/guides/routing/model-variants/* and
+# https://openrouter.ai/docs/faq — dynamic variants :online/:nitro/:floor/:exacto
+# plus static :free and the additional variants :extended/:thinking.
+_OPENROUTER_ROUTING_SUFFIXES: frozenset = frozenset({
+    "free", "extended", "thinking", "nitro", "floor", "exacto", "online",
+})
+
+
+def _strip_openrouter_routing_suffix(model: str) -> str:
+    """Strip a trailing OpenRouter routing suffix if present.
+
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed `agent/model_metadata.py` + `agent/models_dev.py` (+ tests) for PR #97839. The fix is functionally correct: routing suffixes (`:floor`, `:free`, …) are stripped before the catalog lookup so `z-ai/glm-5.3-flash:floor` resolves to the bare models.dev entry (1,310,720 instead of the hardcoded 202,752), while the full routed ID is preserved for provider requests. The candidate-list approach (bare first, stripped second) keeps the original ID authoritative. Tests cover the `get_model_context_length` path, all seven suffixes, provider-scoping, and `get_model_info` ID preservation.

Non-blocking observations:

1. **This overlaps heavily with #97837**, which fixes the same symptom in `agent/models_dev.py` (same functions: `_find_model_entry`, `lookup_models_dev_context`, `get_model_capabilities`, `get_model_info`) with a similar but differently-structured implementation. Both PRs also duplicate the routing-suffix set. Whichever lands first will conflict; the maintainers should reconcile into one implementation and one shared constant (e.g. import the set from one module rather than the "same list as agent/models_dev.py" comment in `model_metadata.py`, which will silently drift if the lists diverge).

2. **`_should_strip_openrouter_suffix` strips for ANY model when `provider is None`** (the permissive branch "so bare-model tests pass"). That changes behavior for existing direct `_find_model_entry(models, model)` callers that pass no provider: a non-OpenRouter provider whose catalog legitimately contains a model literally ending in `:free`/`:thinking`/etc. would now be aliased to the bare entry. In practice model IDs with those suffixes are OpenRouter-specific, so risk is low, but the permissive-None default is a behavior change worth a deliberate look (it exists specifically for tests).

3. **`get_model_context_length` has some redundancy:** the suffix-strip lookup is attempted twice (once in the `effective_provider == "openrouter"` block, once in a second case-insensitive fallback block). It works, but a single normalized lookup helper would be tighter and avoid the near-duplicate loops.

4. **`_is_openrouter_provider_name` is defined in `model_metadata.py` but not referenced in the diff** — if it's dead code, drop it (or wire it into the `effective_provider == "openrouter"` check for consistency with `models_dev`).

Good, well-tested fix; the actionable items are the #97837 reconciliation and the `provider is None` permissive strip.

teknium1:
Closing: #97820 is fixed on main via #108666 (1021a0325696, salvaged from #100651 by @jakobdylanc). The chosen implementation strips only `:nitro`/`:floor`/`:exacto`/`:online`; `:free`/`:batch` are real catalog SKUs with their own context windows (13 differ from their base, e.g. `z-ai/glm-5.2:free` 256K vs 1.05M), so stripping them would over-report the window. Thanks for the report and the fix here.

PR #97841, opened 2026-08-29
fix(state): don't fail-close turns on FTS-index-only errors

### LINKED ISSUE #97794: [Bug]: Healthy state.db fail-closes turns as 'structural corruption' when only the FTS5 index layer errors (false alarm, dropped messages, destructive recovery advice)
## Bug Description

A turn was **killed as `session_persistence_failed` with cause `corrupt`** while `state.db` was provably healthy. The gateway refused to run the turn, told the user the transcript "would have been lost on restart", and recommended `hermes doctor --fix`, `sqlite3 .recover`, and backup restore — all unnecessary and potentially destructive on a healthy database.

The trigger was an **FTS5-index-only failure** ("file is not a database" / "database disk image is malformed" from the FTS shadow tables), which `classify_persistence_error` buckets as whole-file structural corruption and fail-closes the turn.

## Steps to Reproduce

1. Run a gateway (Discord) with a long-lived `state.db` (WAL mode, ~187 MB, 22,835 messages) that has an intermittently failing FTS5 layer (see logs below — this box has had recurring `messages_fts` errors for days).
2. Send a message that triggers a session read/write while the FTS layer is in its failing state.
3. The turn dies with the "⚠️ No reply" message: *"the turn was stopped because the state database reported structural corruption (the transcript would have been lost on restart)... Recovery options: 1. Run `hermes doctor --fix` 2. Salvage with: sqlite3 ~/.hermes/state.db ".recover" 3. Restore from a backup..."*
4. The user's message is dropped and must be re-sent.

## Expected Behavior

- A failure confined to the FTS5 shadow tables must **not** be reported as structural corruption of the whole database, and must not fail-closed the turn.
- The codebase already has a degrade path ("Search temporarily uses LIKE until a later SessionDB open rebuilds the indexes") — FTS-only failures should fail open to that, not kill the turn.
- The "corrupt" guidance (`.recover`, restore backups) should only be shown after verifying the canonical tables are actually damaged (e.g. `PRAGMA integrity_check` mapping damaged b-trees to `sqlite_master` objects, as #88587 does).

## Actual Behavior

Turn killed, message dropped, user told to run destructive recovery on a healthy DB.

## Evidence (2026-08-29, UTC)

`PRAGMA integrity_check` → `ok`, `quick_check` → `ok`, all 22,835 messages + 132 sessions intact. FTS5 tables query fine from a fresh connection (`messages_fts` MATCH returns rows; `integrity-check` on both `messages_fts` and `messages_fts_trigram` passes).

Gateway/agent log sequence:

```
08:54:43 Turn ended: reason=session_persistence_failed ... response_len=0
08:55:32 WARNING gateway.session: Transcript read failed for session ... (returning empty; downstream must not treat this as data loss): database disk image is malformed
08:55:46 ERROR root: FTS5 search failed: file is not a database
08:58:48 WARNING gateway.session: Transcript read failed ...: file is not a database
08:58:48 WARNING ... compression session recovery failed ... (DatabaseError: file is not a database)
08:58:51 ERROR root: FTS5 search failed: file is not a database  (x3)
```

Prior days show the same FTS flapping (agent.log.3):
- `state.db write failed with an FTS-corruption error (fts5: corruption found reading blob 2061584302081 from table "messages_fts") — attempting one-shot in-place FTS rebuild; canonical message rows are preserved.`
- `Could not detach corrupt FTS indexes; canonical write still cannot proceed: malformed database schema ()`
- `Persisted transcript lagged live cached history ... (possible FTS write corruption)`

## Root Cause Analysis

`hermes_state.py` `classify_persistence_error` → `_DB_CORRUPTION_MARKERS` matches any error string containing `"malformed"` / `"not a database"` / `"database corruption"` and returns `"corrupt"` (run_agent.py:3934 renders the scary message). But those strings also come from **FTS5 shadow-table failures** (SQLITE_NOTADB on `messages_fts*`), which are not structural damage to the canonical `messages` table. The classifier has no way to distinguish "FTS index layer broken" from "whole DB file damaged", so an FTS-only hiccup fail-closes the turn and recommends `.recover`/backup restore.

Related: #90837 (real recurring corruption, same box family), #88587 (doctor mislabels structural corruption as FTS — the inverse direction), #86027 (legacy `messages_fts_trigram` reported malformed by newer SQLite — a plausible FTS culprit here: this box links SQLite 3.53.1 in Python vs 3.37.2 system sqlite3).

## Environment

- Hermes v0.20.4 (2026.8.18), upstream 03537d69, local 3ec56e51 (+5 carried commits)
- Ubuntu 22.04.5 LTS, Python 3.11.15
- Python sqlite3 3.53.1; system sqlite3 3.37.2
- state.db: WAL mode, 187 MB, 22,835 messages, 132 sessions
- Platform: Discord gateway

## Proposed Fix

1. In `classify_persistence_error` (or the caller), when the error string matches corruption markers, verify against the canonical tables before returning `"corrupt"` — e.g. run `PRAGMA integrity_check` and map damaged b-trees to `sqlite_master` objects (pattern from #88587). If only FTS shadow tables are implicated, return a new cause (e.g. `"fts_index"`) that degrades search to LIKE and does **not** kill the turn.
2. `session_search` should fall back to LIKE search on FTS5 failure instead of returning `Search failed: file is not a database` (the FTS-rebuild path already exists in `hermes_state`).
3. The `corrupt` user message should only recommend `.recover`/backup restore after canonical-table damage is confirmed.

## Debug Report

`hermes debug share --local` run; key log lines included above. (Public paste skipped — logs contain personal Discord content.)

Finn763:
Closes #97794

## What Problem This Solves

A turn was killed as `session_persistence_failed` with `cause=corrupt` while `state.db` was healthy (`PRAGMA integrity_check → ok`, 22,835 messages / 132 sessions intact, `messages_fts` queries fine). The trigger was an FTS5-shadow-table–only error (`"file is not a database"` / `"database disk image is malformed"` from `messages_fts`), which `hermes_state.py:classify_persistence_error()` buckets via `_DB_CORRUPTION_MARKERS` as whole-file structural corruption. The gateway then fail-closed the turn, dropped the user's message, and advised destructive `hermes doctor --fix` / `sqlite3 .recover` / backup restore.

The codebase already has a degrade path ("Search temporarily uses LIKE until a later SessionDB open rebuilds the indexes") — FTS-only failures should fail open, not kill the turn.

## Solution

- **`hermes_state.py`**: add `PERSISTENCE_ERROR_CAUSES` entry `fts_index` and `_FTS_INDEX_MARKERS = ("messages_fts", "fts5")`. `classify_persistence_error()` now returns `fts_index` when a corruption marker co-occurs with an FTS marker (or `fts5+corrupt` pattern), otherwise `corrupt`. Only non-FTS corruption still maps to `corrupt`.
- **`hermes_state_search.py`**: `_search_messages_impl()` on `DatabaseError` with FTS markers now `logger.warning` + `_try_runtime_fts_rebuild` + `_search_messages_like_fallback(...)` and returns; only if LIKE also fails does it raise. Canonical-table damage still propagates as `corrupt`.
- **`tools/session_search_tool.py`**: `_discover()` on any `Exception` tries `_search_messages_like_fallback`; success returns real discover hits instead of `Search failed: file is not a database`.
- **`run_agent.py` / `gateway/run.py`**: `fts_index` gets a non-destructive message ("transcript intact, Search temporarily uses LIKE until rebuild, no recovery needed"); `corrupt` keeps the `.recover`/backup guidance.

Keeps the fix minimal — distinguishes FTS-layer hiccups from real file damage without adding a full `PRAGMA integrity_check → b-tree → sqlite_master` mapping in this PR.

## Evidence

Isolated `repro_97794.py` (no real `state.db` touched):

```
PERSISTENCE_ERROR_CAUSES includes fts_index: OK
PASS FTS notadb with messages_fts -> fts_index
PASS FTS5 corruption blob -> fts_index
PASS malformed + fts hint -> fts_index
PASS genuine malformed -> corrupt
PASS genuine notadb -> corrupt
PASS fts5 corrupt record -> fts_index
corrupt guidance has .recover=True / LIKE=False
fts_index guidance has .recover=False / LIKE=True
```

In-memory `SessionDB` smoke: inject `DatabaseError: file is not a database` on `messages_fts` MATCH, `search_messages("docker")` correctly falls back to LIKE and returns 2 hits; tool-layer `discover` fallback also PASS.

Before fix: all 6 FTS cases → `corrupt` → session_persistence_failed kill. After: 3 FTS → `fts_index`, 3 genuine → `corrupt`.

## Tests

- Existing `pytest -k "state or fts or persistence"` stash before/after shows no new failures; FTS fallback is covered by new repro and the in-memory SessionDB smoke.
- `PERSISTENCE_ERROR_CAUSES` now `('locked','compression','compression_closed','turn_lease','corrupt','fts_index','disk','unknown')`.

## Risk

Low-medium. Only narrows `corrupt` to non-FTS errors; FTS path degrades to LIKE and rebuilds indexes on next open. No change to canonical-table corruption handling.

## Checklist

- [x] FTS-only `file is not a database` / `malformed` → `fts_index`, not `corrupt`
- [x] Genuine `malformed` without FTS markers still → `corrupt`
- [x] `session_search` degrades to LIKE instead of `Search failed`

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -25511,6 +25511,13 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                 "3. Restore from a backup in ~/.hermes/backups/\n"
                 "Run `hermes doctor` for sanitized diagnostics."
             )
+        elif cause == "fts_index":
+            message = (
+                "⚠️ Session database full-text index degraded — messages are "
+                "still persisted and safe. Search temporarily uses LIKE until "
+                "a later restart rebuilds the indexes. No recovery action is "
+                "needed. Run `hermes doctor` if this persists."
+            )
         else:
             message = (
                 f"⚠️ Session database unavailable — messages may not be persisted. "
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -1855,6 +1855,7 @@ PERSISTENCE_ERROR_CAUSES = (
     "compression_closed",
     "turn_lease",
     "corrupt",
+    "fts_index",
     "disk",
     "unknown",
 )
@@ -1872,6 +1873,18 @@ _DB_CORRUPTION_MARKERS = (
     "file is not a database", # SQLITE_NOTADB (also connection-level poisoning)
     "not a database",
     "database corruption",
+)
+
+# Markers that prove an error is confined to the FTS5 shadow tables.
+# When a corruption marker appears TOGETHER with one of these, the damage
+# is FTS-index-only (e.g. "fts5: corruption found reading blob ... from
+# table \"messages_fts\"" or "database disk image is malformed" while
+# querying messages_fts*).  classify_persistence_error returns "fts_index"
+# for this class so callers degrade to LIKE instead of fail-closing the
+# turn with destructive .recover advice (#97794).
+_FTS_INDEX_MARKERS = (
+    "messages_fts",
+    "fts5",
 )
 
 
@@ -1900,6 +1913,12 @@ def classify_persistence_error(exc_or_str) -> str:
       (``database disk image is malformed`` / SQLITE_NOTADB).  Distinct from
       ``"disk"``: freeing space cannot help, the user needs the repair path
       (``hermes doctor`` / automatic schema surgery).
+    * ``"fts_index"`` — FTS5 shadow-table / index corruption only
+      (``fts5: corruption found ... messages_fts`` or a malformed/notadb
+      error whose message or SQL mentions ``messages_fts*`` / ``fts5``);
+      the canonical ``messages`` / ``sessions`` tables are still healthy,
+      search temporarily degrades to LIKE until a later SessionDB open
+      rebuilds the indexes (#97794).
     * ``"disk"``    — disk full / read-only / permission-shaped failures
       (delegates the disk-full patterns to :func:`is_disk_full_error` so the
       two classifiers can never drift apart — e.g. ENOSPC).
@@ -1925,6 +1944,18 @@ def classify_persistence_error(exc_or_str) -> str:
         return "compression_closed"
     if "being compressed" in text or "compression lease" in text:
         return "compression"
+    # FTS-index-only corruption (issue #97794): when a corruption marker
+    # appears together with an FTS hint (messages_fts / fts5), the damage is
+    # confined to the derived index, not the canonical tables.  Return
+    # "fts_index" so callers degrade search to LIKE and avoid the
+    # destructive .recover guidance that requires canonical B-tree damage.
+    # Also catch "fts5: corruption ..." which may not contain the generic
+    # malformed/notadb markers but still means FTS-only.
+    if any(marker in text for marker in _DB_CORRUPTION_MARKERS):
+        if any(fts in text for fts in _FTS_INDEX_MARKERS):
+            return "fts_index"
+    if any(fts in text for fts in _FTS_INDEX_MARKERS) and "corrupt" in text:
+        return "fts_index"
     # Structural corruption BEFORE the lock and disk buckets: "database disk
     # image is malformed" contains "disk" (and some wrapped corruption
     # strings mention "locked" recovery attempts), so later buckets would
--- a/hermes_state_search.py
+++ b/hermes_state_search.py
@@ -2120,14 +2120,45 @@ class SessionSearchMixin:
                 # FTS5 query syntax error despite sanitization — return empty
                 return []
             except sqlite3.DatabaseError as exc:
-                # A corrupt FTS index raises the malformed / "fts5: corrupt
-                # structure record" class on the MATCH read, the same class the
-                # write path self-heals (#66296). OperationalError (query
-                # syntax) is a subclass caught above; this arm is the corruption
-                # parent. Rebuild the index in place once — the read context
-                # holds no writer lock, so rebuild_fts() can acquire it — and
-                # retry, so search self-heals for read-only sessions (cron/CLI
-                # history search) that never trigger a write to repair it first.
+                # FTS index corruption (#97794): degrade to LIKE instead of
+                # surfacing as whole-DB corruption.  A corrupt FTS MATCH
+                # (malformed / "fts5: corrupt" / SQLITE_NOTADB) still leaves
+                # the canonical messages table healthy — the LIKE scan proves it.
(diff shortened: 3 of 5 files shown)


Enough1122:

Reviewed `hermes_state.py` / `hermes_state_search.py` / `run_agent.py` / `gateway/run.py` / `tools/session_search_tool.py` (+ the new `"fts_index"` cause) for PR #97841. The design is sound and fail-safe: an FTS5-shadow-table-only error is reclassified as `fts_index` so the turn degrades to LIKE search instead of fail-closing with the destructive `.recover` guidance that applies to canonical-table corruption. The user-facing messages in `run_agent.py` and `gateway/run.py` are appropriately reassuring ("messages are still persisted and safe"). I confirmed `_search_messages_like_fallback` exists at `hermes_state_search.py:1541`, so the fallback wiring is valid.

Non-blocking observations:

1. **Message-based classification has an inherent false-positive risk.** The new branch returns `fts_index` whenever a corruption marker AND an FTS marker (`messages_fts`/`fts5`) co-occur — or more broadly when an FTS marker plus the word "corrupt" appear. A genuinely canonical corruption that happens to mention the FTS table in a traceback/context (e.g. during an FTS rebuild that itself surfaced the real damage) could be mislabeled `fts_index`, which then steers the user away from the real repair path. The downside is fail-safe (degrade to LIKE, keep the turn), so this is acceptable — but the second, looser branch (`any fts in text and "corrupt" in text`) is the most permissive and worth a closer look for false positives.

2. **`tools/session_search_tool.py` reaches into a private method via `hasattr(db, "_search_messages_like_fallback")`.** The `hasattr` guard is defensive, but coupling a tool to a private method name means a rename silently disables the fallback (falls through to `tool_error`). Consider exposing this as a public method or checking the class contract, so the fallback can't silently rot.

3. **Minor:** in `_search_messages_impl`, `_try_runtime_fts_rebuild(exc)` is attempted *before* the LIKE fallback even though the index is known corrupt; if it fails (likely), the exception is swallowed and LIKE runs anyway. Harmless (it's a best-effort self-heal), just slightly wasted work in the common corrupt case.

The change correctly implements a real degradation path and keeps user guidance non-destructive. Nothing blocking.

teknium1:
Landed on main in #108130 (04dd80a977). Carried as Co-authored-by under the #97843 salvage; thank you @Finn763.

PR #97842, opened 2026-08-29
fix(update): regenerate dangling editable finder after interrupted install on Windows

### LINKED ISSUE #97819: [Bug][Windows] Interrupted update leaves editable finder mapping pointing at deleted site-packages copy — hermes.exe dies with ModuleNotFoundError
## Summary

On Windows, an interrupted `hermes update` can leave the editable-install finder mapping pointing at a **deleted** site-packages physical copy of the packages. Every launcher then dies with `ModuleNotFoundError: No module named 'hermes_cli'`, and the automated recovery path (`.update-incomplete` retries) never recomputes the mapping.

## Environment

- Windows 11, Hermes Agent v0.20.6 (2026.8.27), `Install method: git`
- HERMES_HOME: `C:\Users\Administrator\AppData\Local\hermes`
- Repo: `%HERMES_HOME%\hermes-agent` (full git clone, editable install via `pip install -e .`)
- Launchers: `%HERMES_HOME%\bin\hermes.exe` and `venv\Scripts\hermes.exe`

## What happened

1. `hermes update` at 17:15 hit Windows' "file in use" (os error 32) — the running launcher exe cannot be replaced. It deferred the dependency reinstall and wrote `venv\Lib\site-packages\.update-incomplete` with `{"attempts": 3}` — already at the recovery-path retry cap, so every startup printed "Pending interrupted-update install has already failed 3 times" and stopped auto-retrying.
2. The deferred install ran later (17:21 / 17:27). It **uninstalled** old physical copies of the packages that a previous install had placed directly in site-packages (`hermes_cli`, `gateway`, `tools`, ...).
3. But the editable finder (`venv\Lib\site-packages\__editable___hermes_agent_0_20_6_finder.py`) still mapped those top-level packages to the now-deleted site-packages paths:

   ```python
   'hermes_cli': 'C:\\...\\venv\\Lib\\site-packages\\hermes_cli'   # <- directory no longer exists
   ```

4. Result: `hermes.exe --version` / `--help` / `update --check` all crashed with `ModuleNotFoundError: No module named 'hermes_cli'`. `update --check` additionally mis-reported "Not a git repository" (it treated site-packages as the project root).
5. Masking gotcha: `venv\Scripts\python.exe -m hermes_cli.main` looked **fine** when run with cwd = repo root, because cwd happened to hit the module — hiding the broken install.

## Root cause

- The editable finder's `MAPPING` can reference site-packages paths (a stale physical-copy location) instead of the repo source tree. When the deferred reinstall removes those physical copies, the mapping dangles.
- The interrupted-update recovery path only retries the dependency install; it never recomputes/regenerates the editable finder mapping, so the launcher stays dead after the retries are exhausted.

## Why this is distinct from existing issues

- #83529 ("hermes update - destroys hermes") — Linux; `uv pip install` aborts on `Permission denied (os error 13)` removing `hermes_agent-0.20.0.dist-info/INSTALLER`; the launcher file itself is gone. Different failure and root cause (uv uninstall permission), not a dangling editable mapping.
- #70211 ("update corrupts editable-install working tree") — npm/package.json corruption from stash/pull/npm on a dirty tree; unrelated to the Python editable finder.

## Recovery (what fixed it)

```bat
cd /d C:\Users\Administrator\AppData\Local\hermes\hermes-agent
venv\Scripts\python.exe -m pip install -e . --no-deps --no-build-isolation
```

This regenerated the finder `MAPPING` against the repo source (verified: every entry now points at `...\hermes-agent\<pkg>`), cleared the `.update-incomplete` marker, and all three launchers work again.

## Suggested fixes

1. When (re)installing editable, detect leftover physical copies of the same packages in site-packages and either remove them or generate the `MAPPING` against the repo source — never against site-packages.
2. The interrupted-update recovery path should verify each `MAPPING` target exists and regenerate the editable install if any target is missing, instead of only retrying the dependency install.
3. `update --check` should anchor repo detection at the install directory (the same one `hermes --version` prints as "Install directory") rather than cwd or site-packages.

---
*Filed from a live Windows install that hit exactly this failure; root cause captured from the broken finder file before repair, recovered as described above.*

Finn763:
Closes #97819

## What Problem This Solves

On Windows an interrupted `hermes update` leaves the editable-install finder (`__editable___hermes_agent_*_finder.py` `MAPPING`) pointing at a deleted `site-packages/hermes_cli` physical copy. `hermes.exe` then dies with `ModuleNotFoundError: No module named 'hermes_cli'`, and the `.update-incomplete` retry path (capped at 3) never regenerates the mapping, so the launcher stays dead. `update --check` also mis-reports "Not a git repository" because it probes `cwd`.

Root: a stale physical copy in `site-packages` shadows the repo source, and recovery only retries the dep install.

## Solution

- **`hermes_cli/_install_repair.py`**: add `_HERMES_EDITABLE_FALLBACK_TOP_LEVELS`, `_get_site_packages_dir`, `_hermes_top_levels_from_pyproject`, `_clean_stale_site_packages_physical_copies` (deletes `site-packages/hermes_cli|gateway|tools|__pycache__`), `_parse_finder_mapping` (AST+regex), `_is_path_inside`, `editable_finder_is_dangling` (missing / maps to site-packages / maps outside repo), `_regenerate_editable_finder` (`pip install -e . --no-deps --no-build-isolation` + quarantine), `ensure_editable_finder_health`. `run_core_install` now cleans physical copies first and validates/regenerates the finder both on success and fallback.
- **`hermes_cli/_early_recovery.py`**: `_complete_pending_core_install` does a lightweight dangling-finder regeneration before counting against the 3-try ceiling, and allows one ceiling-breakthrough `run_core_install` if lightweight fails; re-validates after install.
- **`hermes_cli/main.py`**: `_recover_core_update_marker_locked` validates `editable_finder_is_dangling` after `run_core_install`, cleans and regenerates, and keeps `.update-incomplete` if still dangling.
- **`hermes_cli/update_cmd.py`**: `update --check` now resolves repo via `_resolve_repo_dir()` — tries `PROJECT_ROOT`, `_startup_fast.project_root_str()` (the Install directory printed by `hermes --version`), `get_default_hermes_root()/hermes-agent`, `get_hermes_home()/hermes-agent` — instead of `cwd`; error messages include Install directory and hint `pip install -e . --no-deps`.

All validation happens in temp dirs in tests; never touches real `HERMES_HOME`.

## Evidence

Isolated repro `repro_97819.py` (8 checks, all PASS):

```
[test_parse] PASS
[test_dangling_missing_path] PASS — missing path detected
[test_maps_to_site_packages] PASS — site-packages mapping detected
[test_healthy] PASS — healthy not flagged
[test_clean] PASS — stale copies removed (gateway, hermes_bootstrap.py, hermes_cli, tools)
[test_no_finder] PASS
[test_ensure_health_dangling] PASS — triggers regen for dangling
[test_ensure_health_healthy] PASS — no regen when healthy
```

Manual trace of broken finder before repair showed `hermes_cli: 'C:\...\venv\Lib\site-packages\hermes_cli'` (directory deleted); after `pip install -e . --no-deps` every `MAPPING` entry points at `.../hermes-agent/<pkg>` and launchers work.

## Tests

- `pytest` targeted `update`/`early_recovery` + `py_compile` on 4 changed files → COMPILE_OK, `test_early_recovery_module_is_stdlib_only` PASS, `test_core_marker_retry_ceiling_hands_off_to_late_recovery` PASS, `test_verify_core_dependencies` 31/31 PASS.
- Stash before/after shows no new failures beyond pre-existing Windows `EBUSY` teardown flakes.

## Risk

Medium-low. Windows-specific recovery path; Linux/macOS unaffected. No global constant changes. Cleaning is scoped to known Hermes top-levels from `pyproject.toml`.

## Checklist

- [x] Dangling detection covers missing / site-packages / outside-repo
- [x] Retry ceiling not burned by lightweight regen
- [x] `update --check` anchored to Install directory

### DIFF
--- a/hermes_cli/_early_recovery.py
+++ b/hermes_cli/_early_recovery.py
@@ -616,14 +616,59 @@ def _complete_pending_core_install(root: Path, core_marker: Path) -> bool:
         except OSError:
             attempts = 0
 
-        if attempts >= _EARLY_CORE_INSTALL_MAX_ATTEMPTS:
+        # #97819: dangling editable finder (MAPPING points at deleted
+        # site-packages copy) must be healed even when the full-install
+        # attempt budget is exhausted.  Try a cheap ``pip install -e . --no-deps``
+        # regeneration first without consuming an attempt; if that fails
+        # allow one more full reinstall despite the ceiling.
+        _finder_dangling = False
+        _finder_reasons: list[str] = []
+        try:
+            _finder_dangling, _finder_reasons = ir.editable_finder_is_dangling(root)
+        except Exception:
+            pass
+        if _finder_dangling:
             print(
-                "⚠ Pending interrupted-update install has already failed "
-                f"{attempts} times in the early pass — leaving it for the "
-                "post-import recovery path.",
+                f"⚠ Editable finder MAPPING is dangling: {', '.join(_finder_reasons[:3])} — regenerating editable install...",
                 file=sys.stderr,
             )
-            return False
+            try:
+                ir._clean_stale_site_packages_physical_copies(root)
+            except Exception:
+                pass
+            try:
+                if ir._regenerate_editable_finder(root):
+                    _still, _reasons2 = ir.editable_finder_is_dangling(root)
+                    if not _still:
+                        print("  ✓ Editable finder regenerated in early pass.", file=sys.stderr)
+                        try:
+                            core_marker.unlink()
+                        except OSError:
+                            pass
+                        return True
+                    print(
+                        f"  ⚠ Finder still dangling after lightweight regen: {', '.join(_reasons2[:2])} — will retry full install...",
+                        file=sys.stderr,
+                    )
+                else:
+                    print("  ⚠ Lightweight finder regeneration failed — will retry full install...", file=sys.stderr)
+            except Exception:
+                pass
+            if attempts >= _EARLY_CORE_INSTALL_MAX_ATTEMPTS:
+                print(
+                    "  → Dangling finder detected — retrying full install despite attempt ceiling...",
+                    file=sys.stderr,
+                )
+            # fall through to full install even at ceiling
+        else:
+            if attempts >= _EARLY_CORE_INSTALL_MAX_ATTEMPTS:
+                print(
+                    "⚠ Pending interrupted-update install has already failed "
+                    f"{attempts} times in the early pass — leaving it for the "
+                    "post-import recovery path.",
+                    file=sys.stderr,
+                )
+                return False
 
         if not _claim_recovery_lock(root):
             return False
@@ -636,6 +681,26 @@ def _complete_pending_core_install(root: Path, core_marker: Path) -> bool:
                 file=sys.stderr,
             )
             ir.run_core_install(root)
+            # #97819: post-install finder health sweep — a stale physical
+            # copy could have poisoned the just-finished reinstall.
+            try:
+                _is_dangling, _reasons = ir.editable_finder_is_dangling(root)
+                if _is_dangling:
+                    print(
+                        f"  ⚠ Editable finder still dangling after full reinstall: {', '.join(_reasons[:2])} — regenerating...",
+                        file=sys.stderr,
+                    )
+                    ir._clean_stale_site_packages_physical_copies(root)
+                    if ir._regenerate_editable_finder(root):
+                        _is_dangling2, _reasons2 = ir.editable_finder_is_dangling(root)
+                        if _is_dangling2:
+                            raise RuntimeError(f"editable finder still dangling: {_reasons2}")
+                    else:
+                        raise RuntimeError(f"finder regeneration failed after dangling: {_reasons}")
+            except RuntimeError:
+                raise
+            except Exception:
+                pass
         except Exception as exc:
             new_attempts = ir.bump_marker_attempts(core_marker)
             print(
--- a/hermes_cli/_install_repair.py
+++ b/hermes_cli/_install_repair.py
@@ -28,12 +28,295 @@ import shutil
 import subprocess
 import sys
 import time
+import pathlib
 from pathlib import Path
 
 # Single source of truth for the recovery-lock lifecycle and uv lookup —
 # _early_recovery already owns both, and importing it is free (stdlib-only).
 from hermes_cli import _early_recovery as _er
 
+# ---------------------------------------------------------------------------
+# Editable finder health & stale physical-copy cleanup  (#97819)
+# ---------------------------------------------------------------------------
+_HERMES_EDITABLE_FALLBACK_TOP_LEVELS: frozenset[str] = frozenset(
+    {
+        "hermes_cli",
+        "gateway",
+        "tools",
+        "agent",
+        "cron",
+        "acp_adapter",
+        "plugins",
+        "providers",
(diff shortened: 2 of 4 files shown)


Enough1122:

Reviewed `hermes_cli/_install_repair.py` / `_early_recovery.py` / `main.py` / `update_cmd.py` for PR #97842. The diagnosis is correct and the failure mode is nasty: an interrupted `hermes update` on Windows can leave the editable-install finder's `MAPPING` pointing at a deleted `site-packages/hermes_cli` physical copy, so `hermes.exe` dies with `ModuleNotFoundError` and the `.update-incomplete` retry never regenerates the mapping. The change adds finder-health checks + regeneration across the early-recovery, core-install, and `update --check` paths, and anchors repo detection at the install directory instead of a poisoned `site-packages` path — all reasonable.

Non-blocking observations (several worth a look given the destructive surface):

1. **`_clean_stale_site_packages_physical_copies` deletes by top-level *name* match, including generic names.** The `_HERMES_EDITABLE_FALLBACK_TOP_LEVELS` set includes `"tools"`, `"utils"`, `"cli"`, `"plugins"`, `"gateway"`, `"agent"`, `"cron"` — all extremely common third-party top-level package names. For any non-symlink directory matching one of these in `site-packages`, the code runs `shutil.rmtree(target)`. If a user has any *other* package installed that ships a top-level `tools`/`utils`/`cli`/`plugins` module, this repair step can delete it. It only runs during recovery/repair of interrupted installs, which limits blast radius, but name-matching against a hardcoded generic list is risky. Consider gating the deletion on package provenance (e.g. presence of a `hermes-agent*.dist-info`/RECORD that lists the path) before rmtree, or at minimum restrict the fallback list to clearly Hermes-specific names.

2. **`main.py`'s final guard can reference an unbound `_reasons`.** In `_recover_core_update_marker_locked`, the final guard does `_is_dangling, _ = _ir.editable_finder_is_dangling(...)` then `raise RuntimeError(f"editable finder still dangling: {_reasons}")`. `_reasons` is only bound in the *previous* try block; if that earlier call raised before assignment, the `raise` becomes a `NameError`, which is caught by `except Exception: pass` — silently defeating the guard (marker cleared despite a dangling finder). The internal implementation rarely raises, so this is an edge case, but binding `_reasons` in the guard itself (or using the second tuple element) removes the coupling.

3. **`ensure_editable_finder_health` appears to have no callers** (no references in the current tree or in the diff's hunks). If it's intended for future use, that's fine, but per the repo's dead-code guidance it's worth either wiring it in or dropping it until a consumer exists.

4. **The finder-health + regenerate pattern is duplicated across four files** (`_early_recovery.py`, `main.py`, and twice in `_install_repair.run_core_install`). Centralizing on `ensure_editable_finder_health` (or a single "verify → clean → regenerate" helper) would remove the copy-paste and the inconsistency risk (the four sites use slightly different messaging/behavior).

5. **`editable_finder_is_dangling` returns healthy when no finder files exist.** Absence of `__editable__*finder.py` is not treated as dangling, but a missing finder is equally fatal to the launcher. If the goal is full launcher health, "no finder" should probably be reported, or at least documented as out of scope.

The PR addresses a real, painful failure mode and the fallback-chain approach in `update --check` is thoughtful; the main risk to address is the name-based deletion in item 1.

Issue #90837, opened 2026-08-20
Recurring state.db corruption under gateway-only writes — 11 incidents with hour-resolution onset forensics; every external cause eliminated live


Jay-Verdex:
## Summary

A production gateway (v2026.8.3-era through current main + local containment patches) has corrupted its state.db **11 times between 2026-08-02 and 2026-08-20**, at times reaching daily cadence. We built an hourly onset sentinel that freezes the DB + WAL/SHM + lsof + process table at the first ok→corrupt transition, giving hour-resolution forensics for incidents #5–#11. Over three weeks we eliminated every non-gateway cause **live, one per incident**, and the final two incidents occurred with a **provably gateway-only writer set**.

## Eliminated (each verified in production, not theorized)

| Suspect | How eliminated |
|---|---|
| PYTHONMALLOC/malloc_trim interplay | removed; corruption continued |
| In-process native STT (CTranslate2) | moved out of process; continued |
| Inline FTS trigger write amplification | `fts_mode: deferred`, then FTS **fully off** (no shadow tables, no triggers, no indexer); continued (#9–#11) |
| Native CJK FTS extension | disabled; continued |
| Generic-malformed → live FTS rebuild/fail-open loops | patched fail-closed (kept damage from expanding, did not stop onsets) |
| `close()` `wal_checkpoint(TRUNCATE)` under 30+ live connections | patched to PASSIVE; corruption on the PASSIVE build (#8) |
| External opens with distro SQLite 3.45.1 (WAL-reset-vulnerable) | real contributing vector (agent shell commands + ops scripts); fully sealed via PATH shim + script audits; #10 and #11 occurred with **zero external opens in the onset window** (tool-history + lsof proof) |
| Hourly session-hygiene mutator | excluded for this profile; #11 occurred anyway |
| Kernel/disk/filesystem errors | zero I/O errors, ext4 clean, across all incidents |
| SQLite version/threading | venv links 3.53.1, THREADSAFE=1 |

## What remains

Only the gateway process (venv Python, SQLite 3.53.1, WAL, `synchronous=FULL`, mmap off) holding ~9 descriptors (write conn + read pool + async token accounting) on a 0.6–1.6GB DB, high write volume (~12 cron sessions/day plus live traffic).

## The telling signature (#10)

[...]

PR #88604, opened 2026-08-17
fix(doctor): name structural state.db corruption honestly

liuhao1024:
## What does this PR do?

`hermes doctor` reported every write-health-probe failure as "state.db FTS write corruption" and routed to `--fix` / `hermes sessions repair` (FTS rebuilds). When the corruption is **structural** — canonical tables and indexes damaged, as in the reporter's case where `integrity_check` showed broken b-trees on `sessions`, `session_model_usage`, and two `idx_sessions_*` indexes — an FTS rebuild cannot repair anything: `--fix` fails, leaves a full-size `.malformed-backup-<ts>` beside the DB (a snapshot of the same corrupt file), and the follow-up message tells the operator to restore that useless copy. The tool that actually recovers, `hermes sessions recover --allow-partial` (which rebuilt the reporter's 1.7 GB DB with 144,742/144,742 messages intact), was never mentioned (#88587).

This PR classifies the damage before choosing the message: the new `_state_db_has_non_fts_damage()` discriminator re-runs `integrity_check` read-only, maps damaged tree ids through `sqlite_master.rootpage` (and parses "row N missing from index X" names), and checks whether any damaged object falls outside the FTS shadow set. Structural damage gets its own class name, routes to `hermes sessions recover --allow-partial`, and explicitly warns against restoring the malformed backup. FTS-only damage keeps the existing wording and in-place repair path unchanged; damage that cannot be parsed also keeps the existing path (fail-closed toward the old behavior — never worse than before, per the issue's evidence that misclassification is the harmful direction).

## Related Issue

Fixes #88587

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `hermes_cli/doctor.py`: added `_FTS_OBJECT_RE` / `_TREE_RE` / `_MISSING_INDEX_RE` and the pure classifier `_integrity_damage_is_non_fts(integrity_rows, master_rows)` plus the read-only probe wrapper `_state_db_has_non_fts_damage(state_db_path)`; the write-health failure branch now branches on it — structural damage reports its own class, recommends `sessions recover --allow-partial`, warns against the backup copy, and skips the useless FTS repair attempt (no more 1.7 GB garbage backup)
- `tests/hermes_cli/test_doctor_structural_corruption.py`: 12 regression tests

## How to Test

1. `python -m pytest tests/hermes_cli/test_doctor_structural_corruption.py -q` — Observed result: 12 passed.
[...]

Issue #88587, opened 2026-08-17
doctor reports structural state.db corruption as "FTS write corruption"; --fix cannot repair it and recommends restoring a corrupt backup


shivanathd:
## Summary

When `state.db` corruption is **structural** (canonical tables and their indexes), `hermes doctor`
still reports it as *"state.db FTS write corruption"* and directs the operator to `hermes doctor --fix`
/ `hermes sessions repair`. Those rebuild the FTS index, so they cannot repair the damage, and `--fix`
fails. The message that follows tells the operator to "restore from the backup copy beside state.db",
which is itself a copy of the corrupt database. `hermes sessions recover` — which does fix it — is
never mentioned.

`hermes_state.py` already distinguishes this case; `doctor` does not consult that classification.

## Environment

- Hermes Agent **v0.20.2 (2026.8.16)**, commit `eb63c254bc` (2026-08-17)
- Ubuntu 24.04.4 LTS, Python 3.12.13, sqlite3 3.45.1
- Headless VPS; single gateway plus ~300 cron jobs sharing one `state.db` (~1.7 GB, WAL, `synchronous=FULL`)

## Reproduction / evidence

`PRAGMA integrity_check` reported four corrupt b-trees. Mapping the tree ids through
`sqlite_master.rootpage` shows **none of them are FTS objects**:

```
$ sqlite3 "file:$HERMES_HOME/state.db?mode=ro" "PRAGMA integrity_check(60);"
*** in database main ***
Tree 5  page 421385: btreeInitPage() returns error code 11
Tree 45 page 7701:   btreeInitPage() returns error code 11
Tree 44 page 7739:   btreeInitPage() returns error code 11
Tree 9  page 421465: btreeInitPage() returns error code 11
row 1 missing from index sqlite_autoindex_delivery_obligations_1
...

$ sqlite3 ... "SELECT rootpage, type, name FROM sqlite_master WHERE rootpage IN (5,9,44,45);"
5   table  sessions
9   table  session_model_usage
44  index  idx_sessions_session_key
45  index  idx_sessions_gateway_peer
```

`hermes doctor` nonetheless reported:

```
2. state.db FTS write corruption — run 'hermes doctor --fix' (or 'hermes sessions repair')
   to rebuild the FTS index
```

and `hermes doctor --fix` then failed:

```
Fixed 1 issue(s). 4 issue(s) require manual intervention.
1. state.db FTS write corruption and auto-repair failed — restore from the backup copy
[...]

PR #97843, opened 2026-08-29
fix(state): fail open on FTS5-index-only corruption instead of killing turns

### LINKED ISSUE #97794: [Bug]: Healthy state.db fail-closes turns as 'structural corruption' when only the FTS5 index layer errors (false alarm, dropped messages, destructive recovery advice)

SulthanZahran1:
## What does this PR do?

A healthy `state.db` was reported to the user as **structural corruption** with `.recover` / restore-a-backup advice because `classify_persistence_error` only had one bucket, `corrupt`, for every corruption marker — including the ones FTS5 raises about its own shadow tables (#97794). The reporter's evidence: `PRAGMA integrity_check` → `ok`, 22,835 messages intact, and a turn killed with the destructive banner anyway.

Since #99652 (salvaged #96038 + #97940), current main already **fails open on the write path** for provenance-proven FTS corruption (detach the derived indexes, retry the canonical write, the turn proceeds) and **quarantines** the handle on unscoped corruption. What was still wrong was the *cause the user is shown afterwards*, and that is what this PR fixes. It is rebuilt on current main; nothing from the original head survives except the intent.

1. **One shared provenance rule** — `hermes_state_errors.is_fts_scoped_corruption_error`. A known result code outranks prose: `SQLITE_CORRUPT_VTAB` is FTS-scoped; bare `SQLITE_CORRUPT` / `SQLITE_NOTADB` are unscoped; any other known code (e.g. `SQLITE_CONSTRAINT_TRIGGER`) contradicts FTS-looking text and fails closed. Only without a code (Python < 3.11, RPC-wrapped strings) does the text decide, and then only an `fts5:` corruption report or a corruption marker that names a `messages_fts*` object. `SessionDB._is_fts_write_corruption_error` — and through it the gateway transcript retry — now delegates to it, so the write-repair gate and the persistence classifier can no longer drift (review P1 #2).
2. **New cause `fts_index`** — returned from that rule before any phrase match. An FTS-scoped error that still escapes the write path (the detach was refused) is never rendered as file damage. On main today it classifies as `unknown` (fts5 text carries no marker) or `corrupt` (generic text + `SQLITE_CORRUPT_VTAB` on older SQLite).
3. **New cause `corrupt_unconfirmed`** — the canonical-table probe (review tier 3). The quarantine error now carries the store's `db_path`; when its result code is unscoped, the classifier runs a read-only `PRAGMA quick_check` on a fresh connection and refines the cause only if every reported problem names a `messages_fts*` object (or there is none). **The quarantine decision is untouched** (#97940 / #90837 — the handle still stops writing, the batch is still diverted to JSONL); only the guidance changes from *recover / restore* to *restart, do not recover unless doctor confirms damage*. Any probe failure keeps `corrupt`. `quick_check` rather than `integrity_check`: it walks every b-tree page (the damage class behind a malformed-image error) and skips only the index-content cross-check, which never raises on a write — seconds instead of minutes on a multi-GB store. The verdict is memoised on the error object so the probe runs once per failure, not once per consumer.
4. **Honest wording** at both user-facing boundaries. `fts_index`: *this message was not saved, the store is not damaged, run `hermes doctor --fix` or restart, then send again* (the handle is live). `corrupt_unconfirmed`: *verification found the message tables intact, this process stopped writing as a precaution, restart, unwritten messages were diverted to `sessions/<id>.jsonl`* (the handle is quarantined, so "send again" would be wrong). The gateway startup broadcast gets the same non-destructive text for both.

### What this PR deliberately does not do

- It does not weaken the quarantine on unscoped corruption, and it does not run the probe inside `_execute_write`'s decision — that would have flipped six maintainer tests that pin "bare `SQLITE_CORRUPT` on a live handle quarantines" and reopened #97940.
- It does not add search-path LIKE degradation or `db_path` annotation on `append_message`: main already degrades search to LIKE after fail-open (`_match_rows` / `_search_messages_impl`), and the original annotation would have been dead code behind the quarantine's typed error.
- Reconciling #97841 (@Finn763): same target, string-only classifier (`messages_fts` / `fts5` substrings before any result code), which #96038 showed is unsafe. Its tool-level `_discover` LIKE fallback is not carried over on purpose: swallowing *every* exception there would mask structural corruption as "no results". No unique behaviour or test was carried across, so there is nothing to attribute; I'd suggest closing #97841 in favour of this one.

## Related Issue

Fixes #97794

## Type of Change

## Changes Made

- `hermes_state_errors.py`: `is_fts_scoped_corruption_error`, `verify_canonical_tables_healthy`, `_refine_corrupt_cause`; `PERSISTENCE_ERROR_CAUSES` += `corrupt_unconfirmed`, `fts_index`; `classify_persistence_error` evidence order: type → provenance → phrase → probe
- `hermes_state_fts.py`: `_is_fts_write_corruption_error` delegates to the shared rule
- `hermes_state.py`: `_corrupt_error` stamps `db_path`; `_is_structural_corruption_error` docstring + explicit tuple (raw errors never carry a path, so quarantine is still decided on the result code alone)
- `agent/turn_explainers.py`, `gateway/run_notifications.py`: wording for the two causes
- `tests/state/test_fts_index_fail_open.py` (new, 4 tests): turn-boundary flush succeeds and the exact user message is durable in `messages` after an FTS-only stomp (review P1 #1's requested regression); escaped FTS error → `fts_index`, no quarantine; bare `SQLITE_NOTADB` on a healthy file → `corrupt_unconfirmed` with quarantine + JSONL divert intact, fail-fast re-raise carries the same verdict; real byte-flip damage to the `messages` root page → `corrupt`
- `tests/run_agent/test_turn_completion_explainer.py` (+3 tests): provenance matrix incl. the contradictory-code case and the same verdict from the write gate; probe tier (memoised, never runs on a contradictory code, conservative on missing path / unreadable file); wording for both causes

## How to Test

1. `scripts/run_tests.sh tests/state/test_fts_index_fail_open.py tests/run_agent/test_turn_completion_explainer.py tests/state/test_fts_runtime_rebuild.py tests/hermes_state/test_state_db_corrupt_quarantine.py -q` — 4 files, 87 passed, 0 failed, 1 skipped.
2. Broader affected suites (state, quarantine, NOTADB fail-closed, gateway session / corrupt fallback / normalize, malformed repair, flush divert, compression adoption, salvage gate, session_search tool): 274 passed, 1 skipped; the 2 failures are `tests/gateway/test_session.py::TestSenderPrefixWithBackfill::*` and fail identically on clean `origin/main`.
3. Sabotage (the 5 source files reverted to `origin/main`, tests kept): 5 failed, 26 passed. The two new tests that also pass on main are deliberate contract pins — `test_turn_flush_survives_fts_only_corruption` (main already fails open for proven FTS corruption; this is the turn-boundary proof the review asked for) and `test_unscoped_corruption_on_damaged_file_stays_corrupt` (the probe must never downgrade real damage).
4. `ruff check` clean on every changed file; `git diff --check` clean.

### Code

- [x] I searched for existing PRs (#97841 duplicate target — reconciled above; #96038 landed via #99652 and this composes on top of it; #88604/#88587 are the doctor-side presentation boundary and stay distinct)
- [x] Affected suites pass; pre-existing failures confirmed on clean main
- [x] Tested on: Linux (CachyOS), Python 3.11.15, SQLite 3.53.1

- [x] N/A — no config keys or schemas changed; the classifier, probe and shared rule carry docstrings explaining the evidence order and the quarantine contract

## For New Skills

N/A

## Screenshots / Logs

```
$ scripts/run_tests.sh tests/state/test_fts_index_fail_open.py tests/run_agent/test_turn_completion_explainer.py tests/state/test_fts_runtime_rebuild.py tests/hermes_state/test_state_db_corrupt_quarantine.py -q
=== Summary: 4 files, 87 tests passed, 0 failed, 1 skipped (100% complete) in 7.2s (32 workers) ===

$ python -m pytest <14 affected files> -q
2 failed, 274 passed, 1 skipped   # both failures pre-existing on origin/main (TestSenderPrefixWithBackfill)

$ # sabotage: git checkout origin/main -- <5 source files>; pytest <2 new/changed test files>
5 failed, 26 passed
```

### DIFF
--- a/agent/turn_explainers.py
+++ b/agent/turn_explainers.py
@@ -126,6 +126,27 @@ _PERSISTENCE_CAUSE_EXPLANATIONS: Dict[str, str] = {
         "3. Restore from a backup in ~/.hermes/backups/\n"
         "Then send your message again."
     ),
+    # SQLite scoped the corruption to the FTS index and the derived indexes could not be
+    # detached, so this write did not land; the message store itself is intact (#97794).
+    "fts_index": (
+        "the turn was stopped because the session search index (FTS5) "
+        "is corrupt and could not be detached, so this message was not "
+        "saved. The message store itself is not damaged: do not run "
+        "recovery tools or restore a backup. Run `hermes doctor --fix` "
+        "(or restart Hermes, which repairs the index on open), then "
+        "send your message again."
+    ),
+    # Unscoped corruption report, but a read-only quick_check found the canonical tables
+    # intact: the handle is still quarantined (restart), the recover/restore advice is not.
+    "corrupt_unconfirmed": (
+        "the turn was stopped because the state database reported a "
+        "corruption error, but a read-only verification pass found the "
+        "message tables intact. This process stopped writing to state.db "
+        "as a precaution: restart Hermes to resume. Unwritten messages "
+        "were diverted to sessions/<session_id>.jsonl. Do not run "
+        "recovery tools or restore a backup unless `hermes doctor` "
+        "confirms damage."
+    ),
     "disk": (
         "the turn was stopped because session storage could not "
         "be written (the transcript would have been lost on "
--- a/contributors/emails/zsulthan9@gmail.com
+++ b/contributors/emails/zsulthan9@gmail.com
@@ -1,0 +1,2 @@
+SulthanZahran1
+# PR #97843 fix(state): FTS5-index-only corruption fail-open
--- a/gateway/run_notifications.py
+++ b/gateway/run_notifications.py
@@ -723,7 +723,8 @@ class GatewayNotificationsMixin:
         if not error:
             return
         from hermes_state import _default_db_path, classify_persistence_error, format_session_db_unavailable
-        if classify_persistence_error(error) == "corrupt":
+        cause = classify_persistence_error(error)
+        if cause == "corrupt":
             # Copy-pasteable, so name the real store (profiles / HERMES_HOME do not live under ~/.hermes).
             db_path = _default_db_path()
             message = (
@@ -740,6 +741,15 @@ class GatewayNotificationsMixin:
                 "vulnerable sqlite3 CLI can corrupt it further\n"
                 "3. Restore from a backup in ~/.hermes/backups/\n"
                 "Run `hermes doctor` for sanitized diagnostics."
+            )
+        elif cause in ("fts_index", "corrupt_unconfirmed"):
+            # Index-scoped (or unconfirmed) corruption: the message tables are not damaged, so
+            # the recover / restore advice above would be destructive on a healthy file (#97794).
+            message = (
+                "⚠️ Session database reported a corruption error confined to the search index "
+                "(FTS5); the message tables are not damaged. Messages may not be persisted until "
+                "it is repaired: run `hermes doctor --fix`, then restart the gateway. Do not run "
+                "recovery tools or restore a backup unless `hermes doctor` confirms damage."
             )
         else:
             message = (
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -1017,17 +1017,24 @@ class SessionDB(
     @classmethod
     def _is_structural_corruption_error(cls, exc: BaseException) -> bool:
         """Bare SQLITE_CORRUPT/NOTADB with no FTS provenance: canonical B-tree/schema/freelist damage,
-        never repairable from the live write path."""
+        never repairable from the live write path. The raw error carries no ``db_path``, so the
+        classifier never runs its canonical-table probe here: quarantine is decided on the
+        result code alone, exactly as before (#97940); the probe only refines the user-facing
+        cause of the quarantine error afterwards (#97794)."""
         return (
             isinstance(exc, sqlite3.DatabaseError)
             and not isinstance(exc, StateDbCorruptError)
             and not cls._is_fts_write_corruption_error(exc)
-            and classify_persistence_error(exc) == "corrupt"
+            and classify_persistence_error(exc) in ("corrupt", "corrupt_unconfirmed")
         )
 
     def _corrupt_error(self, prefix: str = "") -> "StateDbCorruptError":
-        """Build the quarantine error for this handle (message assembled once)."""
-        return StateDbCorruptError(f"{prefix}{_STATE_DB_CORRUPT_MSG} (cause: {self._db_corrupt_reason})")
+        """Build the quarantine error for this handle (message assembled once). ``db_path`` lets
+        ``classify_persistence_error`` verify the report against the file before it renders
+        whole-file recovery advice (#97794)."""
+        err = StateDbCorruptError(f"{prefix}{_STATE_DB_CORRUPT_MSG} (cause: {self._db_corrupt_reason})")
+        err.db_path = str(self.db_path)
+        return err
 
     def _halt_db_corrupt(self, exc: BaseException) -> None:
         """Quarantine this handle and raise; never run in-file repair here."""
--- a/hermes_state_errors.py
+++ b/hermes_state_errors.py
@@ -3,7 +3,9 @@ Shared by hermes_state and its mixins; string predicates match wrapped RPC
 strings as well as live sqlite3 exceptions."""
 
 import errno
+import re
 import sqlite3
+from pathlib import Path
 
 # Malformed schema: ``sqlite_master`` itself is inconsistent (typically a DUPLICATE
 # ``CREATE VIRTUAL TABLE messages_fts`` row). SQLite parses the whole schema while
@@ -74,8 +76,8 @@ def is_disk_full_error(exc: BaseException | str | None) -> bool:
 
 # Every classify_persistence_error bucket; consumers enumerate this tuple.
 PERSISTENCE_ERROR_CAUSES = (
-    "locked", "compression", "compression_closed", "turn_lease", "corrupt", "replaced", "disk",
-    "unknown",
+    "locked", "compression", "compression_closed", "turn_lease", "corrupt", "corrupt_unconfirmed",
+    "fts_index", "replaced", "disk", "unknown",
 )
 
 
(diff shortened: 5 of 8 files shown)


andrexibiza (commented):
Reviewed exact head `cb4333e20b3867f3d9a8ce03ee8d07279a0d8ee2` against base `57746cbb842a2e0e58734623567d25102b91db04`, plus current `main@299c652a66bcc915a2a1e10cd2b648f196ec4bba`. I read the 7-file diff, the new regression tests, #97794 and its evidence, the competing #97841 implementation, the merged #92523 salvage history, open #96038's provenance narrowing, and the adjacent #88587/#88604 doctor-side classification work.

There is good work here: moving destructive recovery advice behind a stronger structural/index distinction is the right direction, the LIKE degradation lives at the `SessionDB.search_messages` ownership boundary rather than being duplicated in every caller, and the new tests are materially stronger than marker-only coverage. I do see two blockers before this can be the `Fixes #97794` owner.

**P1 — this changes the diagnosis, but the exact turn still fail-closes.**

The only transcript-write changes in this head annotate the escaping `sqlite3.Error` with `db_path`; they do not alter the caller's stop semantics. On current main, when `_flush_messages_to_session_db(...)` fails, `agent/conversation_loop.py` records the classified cause and then unconditionally sets `_turn_exit_reason = "session_persistence_failed"`, `final_response = ""`, `failed = True`, and breaks specifically so Hermes does not execute tools from unpersisted state. `agent/turn_finalizer.py` in turn stamps every failed `session_persistence_failed` result as an error regardless of the cause. This PR does not touch either boundary.

The new explainer actually exposes the mismatch: for `fts_index` it still says **"the turn was stopped"** and **"send your message again"**. That is the original dropped-message behavior from #97794, only with safer advice. Meanwhile `gateway/run.py` says **"Messages are persisted normally"**, which cannot be true for the exact append that escaped `_execute_write` and caused the failure.

So the core issue contract — a proven FTS-only failure must not kill the turn / drop the user's message — is not yet closed. Please either:

1. make the proven-FTS-only write path recover/detach/retry through to a durable canonical commit *before* the turn continues, with a regression that proves the same user message is actually present in `messages` and the turn proceeds rather than returning `session_persistence_failed`; or
2. narrow this PR to the honest diagnosis + search-degradation slice, remove the `Fixes #97794` closure claim, and leave the turn-continuation/durability slice open for its real owner.

The second option is still valuable; it just should not launder "less destructive failure messaging" into "fail open" completion.

**P1 — the new classifier bypasses SQLite's stronger result-code provenance.**

`classify_persistence_error()` now returns `fts_index` immediately for any text containing both `fts5` and `corrupt`:

```py
if "fts5" in text and "corrupt" in text:
    return "fts_index"
```

That happens before the new database probe and without consulting `sqlite_errorcode`. #96038 is the other side of this exact shape: it demonstrates why known SQLite result codes must outrank error prose. Its regression constructs FTS-looking text paired with `SQLITE_CONSTRAINT_TRIGGER` and correctly requires fail-closed behavior; only `SQLITE_CORRUPT_VTAB` is accepted as old-SQLite FTS provenance, while bare `SQLITE_CORRUPT` / contradictory known codes are rejected. At this head, that contradictory exception would be reclassified as `fts_index` solely because its string says `fts5 ... corrupt`.

Please make the evidence order explicit and shared across the write-repair gate and the persistence classifier: known result code first (`SQLITE_CORRUPT_VTAB` is FTS-scoped; contradictory known codes fail closed), then a narrow text fallback only when no code is available and the error identifies a known `messages_fts*` object, then the canonical-table probe for generic markers. Add the contradictory-code regression here too. That keeps #97843 complementary to #96038 instead of creating two different definitions of "provably FTS-only" in the same file.

**Interlock / merge topology.** #97841 is a duplicate delivery for the same #97794 closure target, not a stack; both should not land independently. If #97843 becomes the owner, reconcile #97841 explicitly and preserve Finn763's credit for any unique behavior/test that is carried over. #96038 is complementary but file-colliding (`hermes_state.py`) and should compose first or be absorbed with attribution; its provenance rule is upstream of this classifier. #92523 is historical merged context explaining why generic malformed text still enters the FTS repair ladder today, not a substitute for either of these fixes. #88604/#88587 is the adjacent doctor/recovery presentation boundary and should stay distinct from runtime turn authority.

**Exact-object evidence.** The PR reports useful local receipts (`127 passed` focused, `171 passed` broader, plus sabotage coverage), but the exact head's upstream CI/Docker/Nix runs are all `action_required`; CI run `33248440949` has **0 jobs**, so there is not yet an upstream execution receipt for this object. The PR base is also two commits behind current main; those two commits are confined to `website/src/data/userStories.json`, so I found no present file collision, but the eventual candidate still needs current-main recomposition and fresh exact-head CI before landing.

This is a strong repair direction, especially the canonical-vs-derived distinction and the search fallback. Closing the two boundaries above would turn it from safer failure reporting into the actual fail-open persistence contract #97794 asks for. 🚀

SulthanZahran1:
Thanks for the thorough review — both P1s were right, and main moved under this PR in the meantime (#99652 landed #96038's provenance rule plus quarantine, and the state module was split), so I rebuilt the change on current main rather than rebase it. Force-pushed; the head is a clean two-commit stack on `origin/main`.

**P1 — the turn still fail-closed.** On current main the write path already fails open for provenance-proven FTS corruption: `_execute_write` detaches the derived indexes and retries the canonical write, and the turn proceeds. The new `tests/state/test_fts_index_fail_open.py::test_turn_flush_survives_fts_only_corruption` is the regression you asked for at the turn boundary: it binds the real `_flush_messages_to_session_db` onto a live `SessionDB`, stomps `messages_fts_data`, and asserts the flush returns `True`, `_last_persistence_error_cause` stays `None`, and the exact user message is the last row in `messages`. The original `db_path` annotation on `append_message` and the search-path LIKE fallback are gone — main owns both now.

The explainer wording no longer contradicts the runtime state. `fts_index` only renders when the detach was refused and the write genuinely did not land, so it says *this message was not saved, the store is not damaged, send again*. The gateway text no longer claims messages are persisted normally.

**P1 — provenance.** There is now one rule, `hermes_state_errors.is_fts_scoped_corruption_error`, with your evidence order: known result code first (`SQLITE_CORRUPT_VTAB` scoped; bare `SQLITE_CORRUPT` / `SQLITE_NOTADB` unscoped; any other known code fails closed), then the narrow text fallback only without a code and only when the report names a `messages_fts*` object, then the canonical-table probe for generic markers. `SessionDB._is_fts_write_corruption_error` delegates to it (so the gateway transcript retry inherits it), and the contradictory-code regression (`SQLITE_CONSTRAINT_TRIGGER` + fts5 text) is in `test_classify_persistence_error_fts_provenance_order`, asserting the same `False` from both the classifier and the write gate.

**Tier 3 and the quarantine.** I kept the probe out of the quarantine decision on purpose. Running it inside `_execute_write` would have flipped the six tests that pin "bare `SQLITE_CORRUPT` on a live handle quarantines" and reopened #97940. Instead the quarantine error carries `db_path`, and the classifier runs a read-only `quick_check` afterwards; a clean verdict yields a distinct cause, `corrupt_unconfirmed`, whose guidance is *restart, do not recover / restore unless doctor confirms damage* — because the handle **is** quarantined and "send again" would be a lie. Real byte-flip damage keeps `corrupt` (tested). Happy to fold `corrupt_unconfirmed` into `fts_index` if you'd rather have one cause, but the two runtime states need different advice.

**#97841.** Same target, string-only classifier ahead of result codes (the #96038 shape). Nothing unique carried over — its tool-level catch-all LIKE fallback would mask structural corruption as "no results" — so there's no attribution to preserve; suggest closing it in favour of this one.

Receipts are in the PR body (focused files via `scripts/run_tests.sh`, 274 broader tests, sabotage run, ruff/diff-check).

teknium1:
Landed on main in #108130 (04dd80a977). Cherry-picked with authorship intact (Co-authored-by @Finn763 from #97841); thank you @SulthanZahran1. The `quick_check` `corrupt_unconfirmed` tier was dropped: on 3.53.1 `quick_check` raises rather than reports on a torn shadow b-tree, and softening quarantine guidance by probe weakens the fail-close.

PR #79533, opened 2026-08-05
fix(cli): finalize Relay session in one-shot cleanup (#79471)

thatssoheil:
## Summary

One-shot mode (`hermes -z` / `--oneshot`) hard-exits via `os._exit`, so the Relay conversation was never finalized: exported telemetry contains session start + turn lifecycle but **no session end** — the root session stays open (#79471).

`_cleanup_oneshot_runtime()` already mirrors `cli.py:_run_cleanup()` for the other process-global clients (terminal, delegation, browser, MCP, auxiliary), but never finalizes the Relay conversation.

## Fix

- **`hermes_cli/main.py`**: `_cleanup_oneshot_runtime(session_id)` now calls `hermes_cli.lifecycle.finalize_session(session_id=..., platform="cli", reason="oneshot_complete")` — the same chokepoint `cli.py` single-query and the TUI gateway already use — before the hard exit.
- **`hermes_cli/oneshot.py`**: `run_oneshot()` returns `(rc, session_id)` so the session id reaches the cleanup path; `_run_agent`'s result dict already carries it (`turn_finalizer.py` sets `"session_id": agent.session_id`).

## Safety

- Best-effort and guarded (`try/except Exception: pass`), mirroring `cli.py:_notify_session_finalize` (cli.py:1286-1294).
- No Relay host → `finalize_conversation` no-ops via `for_profile(create=False)` → `None`; `close_session` on an unknown session is a no-op. Safe for installs without Relay.
- The `#43055` hard-exit safety boundary is preserved: the `finally` block still runs `os._exit`; a finalization failure never falls through to interpreter finalization.
- `run_oneshot`'s only caller is `_run_and_exit_oneshot` (verified; `tui_gateway/methods_session.py` imports the unrelated `agent.oneshot`).

## Tests

`tests/hermes_cli/test_oneshot_relay_finalize.py`:
- `test_cleanup_oneshot_runtime_finalizes_relay_session` — proves `finalize_session` fires with session_id/platform/reason; all other cleanup clients still called.
- `test_cleanup_oneshot_runtime_without_session_id_skips_finalize` — no session id → no finalize (idempotent no-op).

**RED proof**: with the fix stashed, the test fails (`TypeError: _cleanup_oneshot_runtime() takes 0 positional arguments`). GREEN with the fix: `2 passed`.
Regression: `543 passed` across oneshot/lifecycle/TUI-gateway suites.

## Validation

[...]

Issue #10200, opened 2026-04-15
Bug: auxiliary_client._client_cache accumulates AsyncOpenAI clients indefinitely, causing fd exhaustion in long-running gateway


morningto:
## Summary

`auxiliary_client._client_cache` (line ~1685) accumulates `AsyncOpenAI` client entries indefinitely in long-running gateway processes. Each cached async client holds an `httpx.AsyncClient` bound to a specific event loop, which keeps the loop's kqueue selector and self-pipe unix sockets alive. Over days of operation, this exhausts the process file descriptor limit.

This is a deeper root cause than #8043, which addressed `model_tools.py` event loop cleanup but did not address the `_client_cache` accumulation.

## Root Cause

The cache key for async clients includes `id(asyncio.get_event_loop())` (line ~1817):

```python
cache_key = (provider, async_mode, base_url, api_key, loop_id)
```

When worker threads are recycled by `ThreadPoolExecutor` (e.g., during cron job execution or gateway message handling), new threads get new event loops with new `loop_id` values. Each unique `loop_id` creates a new cache entry with a new `AsyncOpenAI` client. Old entries are never evicted — `cleanup_stale_async_clients()` (line ~1769) only removes entries whose loop `.is_closed()`, but the cached client itself holds a reference to the loop, preventing it from being closed.

## Reproduction

```python
import asyncio, concurrent.futures, gc, os, subprocess
from agent.auxiliary_client import _client_cache, _client_cache_lock, _get_cached_client

pid = os.getpid()

def fd_count():
    r = subprocess.run(["lsof", "-p", str(pid)], capture_output=True, text=True)
    lines = r.stdout.strip().split("\n")[1:]
    kq = sum(1 for l in lines if "KQUEUE" in l)
    return len(lines), kq

print(f"Before: total={fd_count()[0]}, KQUEUE={fd_count()[1]}")

for i in range(10):
    pool = concurrent.futures.ThreadPoolExecutor(max_workers=1)
    def run():
        loop = asyncio.new_event_loop()
        asyncio.set_event_loop(loop)
        loop.run_until_complete(asyncio.sleep(0))
        _get_cached_client("custom", "test", async_mode=True,
                           base_url="https://example.com/v1", api_key="key")
    pool.submit(run).result()
    pool.shutdown(wait=True)

gc.collect()
print(f"After: total={fd_count()[0]}, KQUEUE={fd_count()[1]}")
[...]

PR #97844, opened 2026-08-29
fix(cron): fire on_session_finalize when a cron job finishes

BankzNNBB:
## What does this PR do?

`cron/scheduler.py:run_job()` opens a cron session through `AIAgent` — which
emits `on_session_start` — and closes the SQLite row with `end_session()`, but
never notifies plugin lifecycle observers. Every cron run therefore emitted a
session start with no matching end.

Measured on a live install against a plugin that posts session boundaries to an
external store:

| source | sessions with a matching end |
|---|---|
| cli | 3/3 every day since 24 Aug |
| telegram | 17/26 (the 9 without still had `ended_at = NULL` — genuinely open) |
| **cron** | **0/51** |

The `cli` and `telegram` numbers are what makes this a `run_job` bug rather than
a hook bug: the same plugin fires correctly on every other path. Probing the
hook directly returned `200 /session/end`, so nothing downstream was broken — it
was simply never called.

Why this approach: `end_session()` already exists two lines above and is the
right place for the DB row, but it is a `SessionDB` method with no knowledge of
plugins. `lifecycle.finalize_session()` is the existing seam every other path
uses (`gateway/run.py` on shutdown and expiry, `tui_gateway._finalize_session`),
so this adds the missing call rather than inventing a mechanism.

## Related Issue

Sibling call path to #79533, which fixes the same missing-end symptom in
one-shot cleanup (`os._exit` skipping `atexit`). That PR does not touch
`cron/scheduler.py`, and cron does not route through `run_oneshot` — it has its
own `run_job`. The two are independent and can land in either order.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- **`cron/scheduler.py`** (+27): call `lifecycle.finalize_session()` in
  `run_job`'s session teardown.
  - Passes `_final_cron_session_id`, not the raw `_cron_session_id` captured
    before `AIAgent` started. Compression can rotate the agent onto a
    continuation mid-run, and the stale id would name a session observers
    cannot resolve — `run_job` already computes the resolved id for
    `end_session`, so this reuses it.
  - Ordered before `_session_db.close()` so a hook can still read session state.
  - Wrapped in `try/except (Exception, KeyboardInterrupt)` so a raising plugin
    cannot skip the SQLite teardown below — that teardown is what stops a
    long-lived gateway from leaking one fd per job until EMFILE (#10200).
- **`tests/cron/test_cron_session_finalize.py`** (+180, new): three regression
  tests driving `run_job` end to end.

## How to Test

Reproduction on `main`, with any plugin implementing `on_session_finalize`:

1. Register a plugin that logs `on_session_start` and `on_session_finalize`.
2. Run any agent-backed cron job (`no_agent` jobs never create a session, and a
   `monitor_script` job whose hash is unchanged is suppressed before the agent
   starts — neither will reproduce this).
3. Observe a start with no matching finalize.

Proof the fix works, from the same install after applying it — a real cron run,
not a test double:

```
15:17:44  OK 200 /session/start  cron_0d4c6cbe4057_20260829_151742
15:17:46  OK 200 /session/end    cron_0d4c6cbe4057_20260829_151742
```

Automated:

```bash
pytest tests/cron/test_cron_session_finalize.py -q     # 3 passed
scripts/run_tests.sh tests/cron/ -q                    # 1036 passed, 0 failed
```

Each of the three tests was verified to **fail** with the finalize call removed
from `run_job`, so they cannot pass vacuously:

```
FAILED test_run_job_fires_session_finalize
FAILED test_run_job_finalizes_compression_continuation
FAILED test_finalize_hook_failure_does_not_skip_sqlite_teardown
```

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate — found #79533 on the one-shot path and deliberately scoped this to `cron/scheduler.py` so the two do not conflict
- [x] I've run `pytest tests/ -q` and all tests pass — see note below
- [x] I've tested on my platform: macOS 26.5.2 (Apple silicon), Python 3.11

> Note on the full-suite checkbox, since "all tests pass" needs qualifying:
> the whole tree on this branch is **39590 passed / 203 failed**. To attribute
> those, I re-ran the same 68 failing files against a clean `origin/main`
> worktree with no patch applied: **201 of 203 fail identically there**, so they
> are pre-existing. The 2-test delta was one file,
> `tests/tools/test_mcp_tool_issue_948.py`, which passes 3/3 in isolation on
> *both* the patched branch and clean `main` — flaky under `-j 56`, and an MCP
> test with no path to `cron/scheduler.py`. `tests/cron/` itself is
> **1036 passed / 0 failed**.

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A, the change is an internal call with an explanatory comment
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — N/A, no config keys
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — no platform-specific calls; pure Python control flow

## Screenshots / Logs

Session-boundary counts from the affected install before the fix — 369 starts
against 71 ends, with cron and subagent contributing zero ends:

```
/session/start  369
/session/end     71

by source:  cli 3/3 daily · telegram 17/26 · cron 0/51 · subagent 0/12
```

### DIFF
--- a/cron/scheduler.py
+++ b/cron/scheduler.py
@@ -6880,6 +6880,33 @@ def run_job(
                 )
             except (Exception, KeyboardInterrupt) as e:
                 logger.debug("Job '%s': failed to end session: %s", job_id, e)
+            # Fire on_session_finalize for this cron run.
+            #
+            # end_session() above closes the DB row; it does NOT notify plugin
+            # lifecycle hooks. gateway/run.py finalizes on shutdown and expiry,
+            # tui_gateway in _finalize_session — run_job had no equivalent, so
+            # every cron run emitted a start with no matching end. Measured
+            # 2026-08-29 against a plugin that posts session boundaries: 51/51
+            # cron sessions opened and never closed.
+            #
+            # Uses _final_cron_session_id, not _cron_session_id: compression can
+            # rotate the live agent onto a continuation mid-run, and the raw id
+            # captured before AIAgent started would name a session the hook
+            # cannot resolve. Ordered before close() so a hook may still read
+            # session state, and wrapped so a misbehaving plugin cannot skip the
+            # SQLite teardown below (fd leak — #10200).
+            try:
+                from hermes_cli import lifecycle
+
+                lifecycle.finalize_session(
+                    session_id=_final_cron_session_id,
+                    platform="cron",
+                    reason="cron_complete",
+                )
+            except (Exception, KeyboardInterrupt) as e:
+                logger.debug(
+                    "Job '%s': session finalize hooks failed: %s", job_id, e
+                )
             try:
                 _session_db.close()
             except (Exception, KeyboardInterrupt) as e:
--- a/tests/cron/test_cron_session_finalize.py
+++ b/tests/cron/test_cron_session_finalize.py
@@ -1,0 +1,180 @@
+"""Regression tests: a cron run must notify session lifecycle hooks.
+
+``run_job`` closes its SQLite session row via ``end_session``, but that call
+does not reach plugin lifecycle observers. Before this was fixed, every cron
+run emitted ``on_session_start`` with no matching ``on_session_finalize``:
+measured against a plugin that posts session boundaries to an external store,
+51/51 cron sessions were opened and never closed, while the CLI and gateway
+paths matched every day.
+
+The three properties pinned here:
+
+1. ``run_job`` calls ``lifecycle.finalize_session`` for its cron session.
+2. It passes the *resolved* session id. Compression can rotate the live agent
+   onto a continuation mid-run, so the raw id captured before ``AIAgent``
+   started would name a session the hook cannot resolve.
+3. A plugin that raises cannot skip the SQLite teardown that follows — that
+   teardown is what stops a long-lived gateway from leaking one fd per job
+   until it hits EMFILE.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+import cron.scheduler as cron_scheduler
+from hermes_cli import lifecycle
+
+
+class _RecordingSessionDB:
+    """SessionDB stub that records teardown order for assertions."""
+
+    instances: list["_RecordingSessionDB"] = []
+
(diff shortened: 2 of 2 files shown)


Enough1122:

**fix(cron): fire on_session_finalize when a cron job finishes**

Solid fix with strong regression tests. Firing `lifecycle.finalize_session` after `end_session` and ordering it *before* `close()` (so a hook can still read session state) is the right call, and the wrapped try/except correctly guarantees the SQLite teardown always runs even when a plugin raises. The use of `_final_cron_session_id` (resolved/compression-rotated id) rather than the raw pre-agent id is well-reasoned and pinned by `test_run_job_finalizes_compression_continuation`.

Non-blocking observations:

1. **Debug-level logging of hook failures.** The `except (Exception, KeyboardInterrupt)` around the finalize call only logs at `logger.debug`. That is consistent with the surrounding teardown handlers, but a genuinely misbehaving lifecycle plugin will be effectively invisible in normal (INFO) logs — the fd-leak symptom it protects against will re-appear silently. Consider logging unexpected (non-`KeyboardInterrupt`) exceptions at WARNING while still never skipping teardown.

2. **Placement vs. failed/aborted runs.** The hunk lands after `end_session` in what appears to be the normal completion path. If a cron job fails before reaching that point (exception, interrupt, or the early `return`-style paths), `on_session_finalize` will not fire, so the "every start has a matching end" invariant claimed in the docstring may not hold for failed runs. If the intent is coverage for all terminal states, verify the finalize call also runs on the failure/abort branches (or explicitly document it as completion-only).

3. **Lazy import.** `from hermes_cli import lifecycle` inside the function is a reasonable circular-import guard, but it means a broken lifecycle import is only discovered at cron runtime, not at startup. Minor.

PR #97845, opened 2026-08-29
fix(state): close the SessionDB lock gate's blind spot on its own mixin files

nftpoetrist:
## Summary

`tests/state/test_no_locked_readers_gate.py` (added by #97676, merged earlier in this same window) parses `hermes_state.py`'s `SessionDB` class body with `ast` and flags any method that holds the writer lock (`self._lock`) around a pure-read query — Pattern C, where every concurrent turn's persistence convoys behind an unrelated read. #97676 converted 39 such methods and closed detection gaps for alias/variable-SQL readers.

But `SessionDB` is declared as:
```python
class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin):
```
— and the gate only ever opened `hermes_state.py`. It never parsed the three mixin files those base classes live in (`hermes_state_search.py`, `hermes_state_schema.py`, `hermes_state_portability.py`), so a locked reader declared there was structurally invisible to the scanner regardless of how good the alias/variable-SQL detection got.

Applying the gate's exact scanning logic directly to the three mixin files turns up **9 genuine pure-read methods still holding the writer lock**, none in #97676's converted list:

- `hermes_state_search.py`: `_fts_teardown_trash_step`, `fts_optimize_available`, `optimize_fts_storage`, `list_recent_user_messages`
- `hermes_state_portability.py`: `distinct_session_cwds`, `list_cron_job_runs`, `_get_session_rich_rows_batch`, `list_skill_scaffolded_sessions`, `get_first_assistant_text`

`_get_session_rich_rows_batch` is a hot path — it backs `list_sessions_rich`'s compression-tip resolution and the web server's session-search hydration across every gateway install. Its own docstring already claims "same read-your-writes guarantee as `list_sessions_rich`" — but `list_sessions_rich` was already using `_read_ctx()` (its guarantee comes from `flush_token_counts()` before the read, not from holding the writer lock), while this method's implementation never caught up to match its own sibling's stated contract.

## Fix

1. Converted all 9 methods to `with self._read_ctx() as conn:`, the exact pattern #97676 used for its 39 conversions. Verified each is a genuine pure read with no hidden writes by tracing every helper it calls (`_db_has_legacy_inline_fts`, `_has_fts_trash`, `_fts_external_index_empty_with_messages`, etc.) down to their own SQL.
2. Extended the gate itself (`_ALL_STATE_SOURCES`) to scan all three mixin files under their own class name, in addition to `hermes_state.py`, so this blind spot can't silently reopen.
3. Added a regression test (`test_scan_all_state_sources_visits_every_mixin_file`) that plants a synthetic violation in a mixin-shaped file and asserts the scanner still catches it — a future change that narrows the file list back to one file passes the existing single-file sabotage test but fails this one.

## Test plan

- [x] Independently re-derived the gate's own scanning methodology and applied it to the 3 mixin files by hand before writing any fix, to confirm the 9 violations are real (not a scanner false-positive) — each one's helper calls trace to plain SELECTs, no writes.
- [x] `tests/state/test_no_locked_readers_gate.py` (3 tests, including the new one) passes.
- [x] Mutation-verified: stashed the 9 method conversions (kept the gate's new multi-file scope), confirmed `test_no_locked_pure_readers` fails and names all 9 real violations with correct file/line. Restored the fix, confirmed clean.
- [x] Full `tests/state/` (92 tests) and the broader SessionDB-adjacent suites (`test_hermes_state.py`, `test_session_system_prompt_dedup.py`, `test_session_skill_previews.py`, `test_fts_cjk_bigram.py`, `test_wal_checkpoint_strategy.py`, `test_list_recent_user_messages_handoffs.py`, `tests/tui_gateway/test_session_profile_db.py` — 295 tests total) pass unchanged, 2 pre-existing skips.

### DIFF
--- a/hermes_state_portability.py
+++ b/hermes_state_portability.py
@@ -54,8 +54,8 @@ class SessionPortabilityMixin:
         where = "cwd IS NOT NULL AND TRIM(cwd) != ''"
         if not include_archived:
             where += " AND archived = 0"
-        with self._lock:
-            rows = self._conn.execute(
+        with self._read_ctx() as conn:
+            rows = conn.execute(
                 "SELECT cwd AS cwd, COUNT(*) AS sessions, "
                 "MAX(COALESCE(ended_at, started_at, 0)) AS last_active "
                 f"FROM sessions WHERE {where} GROUP BY cwd"
@@ -119,8 +119,8 @@ class SessionPortabilityMixin:
             ORDER BY s.started_at DESC, s.id DESC
             LIMIT ? OFFSET ?
         """
-        with self._lock:
-            cursor = self._conn.execute(query, (prefix, prefix_hi, limit, offset))
+        with self._read_ctx() as conn:
+            cursor = conn.execute(query, (prefix, prefix_hi, limit, offset))
             rows = cursor.fetchall()
 
         runs: List[Dict[str, Any]] = []
@@ -202,8 +202,8 @@ class SessionPortabilityMixin:
             {prompt_join}
             WHERE s.id IN ({placeholders})
         """
-        with self._lock:
-            cursor = self._conn.execute(query, ids)
+        with self._read_ctx() as conn:
+            cursor = conn.execute(query, ids)
             rows = cursor.fetchall()
         result: Dict[str, Dict[str, Any]] = {}
         for row in rows:
@@ -229,8 +229,8 @@ class SessionPortabilityMixin:
         Returns ``id``, ``title``, and the full first-turn ``content`` so a
         caller can re-derive what the user typed. Newest first.
         """
-        with self._lock:
-            rows = self._conn.execute(
+        with self._read_ctx() as conn:
+            rows = conn.execute(
                 """
                 SELECT s.id, s.title, m.content
                 FROM sessions s
@@ -254,8 +254,8 @@ class SessionPortabilityMixin:
         Pairs with :meth:`list_skill_scaffolded_sessions` so a re-title can feed
         the titler the same (request, reply) shape the live path uses.
         """
-        with self._lock:
-            row = self._conn.execute(
+        with self._read_ctx() as conn:
+            row = conn.execute(
                 "SELECT content FROM messages "
                 "WHERE session_id = ? AND role = 'assistant' AND content IS NOT NULL "
                 "ORDER BY timestamp, id LIMIT 1",
--- a/hermes_state_search.py
+++ b/hermes_state_search.py
@@ -189,9 +189,9 @@ class SessionSearchMixin:
         comparison, so they keep the legacy chunked ``LIMIT`` delete —
         those shadow tables are small by construction.
         """
-        with self._lock:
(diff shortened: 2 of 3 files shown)


Enough1122:

**fix(state): close the SessionDB lock gate's blind spot on its own mixin files**

Good fix on both halves: the reader-writer contention change routes pure-read methods in `hermes_state_portability.py` / `hermes_state_search.py` through `_read_ctx()` instead of holding the writer lock, and the gate test now scans all four files that contribute `SessionDB` methods (`hermes_state.py` + the three mixins). The sabotage self-check that plants a locked reader in a fake mixin file is a good guard against the scope silently narrowing again.

Non-blocking observations:

1. **`_read_ctx()` semantics.** The change substitutes `conn` (from `_read_ctx()`) for `self._conn` in each body. Assuming `_read_ctx()` returns the same connection under a shared/read lock (the established Pattern-C fix), this is correct. If instead it ever returns a *different* read-only connection, methods relying on a write-consistent snapshot or on writes through that handle could silently misbehave — worth a quick confirmation that `_read_ctx()` hands back the same `self._conn` (or a coherent reader) for these queries.

2. **Hardcoded mixin file/class list in the gate.** `_ALL_STATE_SOURCES` hardcodes the three mixin file names and class names. If `SessionDB`'s inheritance changes (new mixin, renamed class), the gate could silently narrow again — exactly the failure mode this PR fixes. Consider deriving the mixin list from `SessionDB.__mro__` (or asserting the hardcoded classes are still in the MRO) so the gate fails loudly when the file set drifts. The added sabotage test helps, but only covers what the list already includes.

kshitijk4poor:
**Triage note (perf/P2 sweep)** — verified, and the gate has teeth.

Applies cleanly to `origin/main` 1cb3ab6173; 4/4. Running the new AST gate (`tests/state/test_no_locked_readers_gate.py`) against **main** flags exactly the blind spot the PR names — 9 pure-read methods holding the writer lock across the mixin files (`_fts_teardown_trash_step`, `fts_optimize_available`, `optimize_fts_storage`, `list_recent_user_messages`, `distinct_session_cwds`, `list_cron_job_runs`, `_get_session_rich_rows_batch`, `list_skill_scaffolded_sessions`, `get_first_assistant_text`); with the PR, 0. Mutation check: reverting only `hermes_state_portability.py` to main re-fails the gate.

Convoy bench (`bench/s2_lock_convoy.py`: 20k sessions, a thread hammering `distinct_session_cwds()` while the main thread does 200 `append_message`s): writer p95 21.8–29.4 ms → 17.8–20.9 ms, max 64–232 ms → 24–31 ms in 2 of 3 runs. Modest on this fixture; the value here is the structural guarantee rather than one number.

kshitijk4poor:
Merged via #101699 (rebase — your commit landed on `main` as c7429f60ca with authorship preserved). Thank you @nftpoetrist — 9 pure readers out from under the write lock, and the gate now covers the mixins so it can't regress silently.

PR #96162, opened 2026-08-27
fix(desktop): enable cross-source group chat creation

jugol:
## Summary

- enable **New Group Chat** when at least two selectable bots exist across the full multi-source roster
- continue excluding unreachable `ghost` placeholders, matching the creation dialog's own selection rules
- add a regression contract for local-plus-remote Bot fleets

## Why

The menu gate counted `activeSourceRoster`, which intentionally excludes `remoteSource` rows. A Desktop with one local Bot and one or more healthy Bots on registered connections therefore disabled **New Group Chat**, even though the dialog already receives the full multi-source `roster` and can route those members to their owning machines.

## Reproduction

1. Run Hermes Desktop with one local profile.
2. Register another reachable gateway that exposes at least one Bot profile.
3. Open **Bots → New…**.
4. Observe **New Group Chat** is disabled even though two selectable Bots exist across the roster.

After this change, the item is enabled whenever the full roster contains at least two non-ghost Bots.

## Tests

- `node --test src/plugins/hermes-bots/tests/create-group-chat.test.mjs` — 5 passed
- `node --test src/plugins/hermes-bots/tests/*.test.mjs` — 605 passed
- `npm run build` in `apps/desktop` — passed
- `git diff --check` — passed

## Manual validation

- macOS 26.5.1 (arm64) Desktop
- one local Bot plus Bots from registered Windows, macOS, and NAS connections
- verified **New Group Chat** enables and the picker can seat cross-machine members

Related to #94726.

PR #97797, opened 2026-08-29
feat(bot-mode): add scoped cross-gateway Group Chat transport

dokterdok:
## The user problem

#97744 lets a Group Chat continue on one gateway after Desktop closes. A Group
Chat containing Bots from two gateways still needs a safe way for the home
gateway to dispatch an exact member turn, recover it after response loss, and
observe Stop or completion without using Desktop as a courier.

This PR adds the scoped cross-gateway Group Chat transport, internally called RoomLink. It is the direct, text-first layer under the later Desktop continuity UX.

> [!IMPORTANT]
> This PR is stacked on #97712 and #97744. Review only the commits after the
> current #97744 head `ee2f98efe1`. It does not enable a new Desktop creation path by itself.

## Cross-gateway transport contract

- Durable `/v1/runs` admission and status using one idempotency identity and
  immutable request fingerprint.
- Asynchronous peer commands that recover the same admitted run after response
  loss or process restart.
- A target-issued RoomLink grant scoped to the exact room, home and authority
  epoch, member, target install, target profile, permissions, execution policy,
  and expiry.
- Room grant authentication happens before request-body parsing or active-work
  reservation. A status-only grant cannot mint new dispatch authority.
- Remote execution requires manual or smart approvals; an execution-policy or
  capability change pauses the route for explicit reauthorization.
- Live capability verification before first dispatch.
- Direct authenticated transport with credential-safe redirect handling.
- Durable route and run receipts, bounded status polling, replay-safe dispatch,
  and exact-run Stop.
- Fail-closed mixed-version behavior: unsupported peers stay unavailable rather
  than weakening the grant or silently running a different Bot.

## Reliability behavior

- A peer proven unreachable before admission keeps the exact turn queued for a
  bounded retry.
- Timeout, reset, or 5xx after submission remains uncertain and reuses the same
  durable dispatch identity; it is never sent through a second route.
- Completion may win a Stop race, but Stop remains pending until the exact
  execution generation is terminal.
- Home restart reloads the route and receipt without resubmitting speculative
  work.
- Message identity is bound to immutable content; conflicting replay fails.
- Superseded authority, stale reservations, revoked grants, and late receipts
  cannot commit new room state.

## Deliberate boundaries

- Direct link only; no mesh, relay, or automatic authority failover.
- Desktop connected to both gateways is not a data path. If only one gateway
  is reachable from every participant, that gateway should hold authority;
  reverse-dial or a relay is a later NAT-traversal increment.
- Cross-gateway text turns only.
[...]

PR #97846, opened 2026-08-29
feat(groups): integrate hosted Group Chats into Desktop

dokterdok:
## Return to a hosted Group Chat without duplicating work

The gateway can keep a Group Chat running after Desktop closes, but that alone does not give Desktop a reliable way to return to it. People need the right group, current history and progress, and commands that do not create duplicate work.

This draft delivers the Layer 4 Desktop client integration on the #106742 runtime at `b36398f71929`. Layer 2 remains the execution owner: Desktop selects the compatible mode, restores the conversation and sends durable commands. Older gateways retain the explicit classic Desktop-driven fallback. Gateway storage, Files, messaging and recovery remain separate owners.

## Prevent stale requests from crossing gateways

**P1 addressed — switching gateways could send an old request to the wrong host.** While Desktop was resolving a shared-remote profile, changing the primary gateway could cause the pending request to use the replacement connection. Rejecting the response afterward was too late: the request had already crossed the wrong connection.

The lookup now retains its exact gateway, connection, profile and ownership generation through resolution and dispatch. Obsolete success, failure and isolated-descriptor results are rejected before sending or opening a fallback socket. Switching away and back cannot revive the old lookup. Request, open, ensure and retain use the same rule; valid current-owner routing and supplied cancellation still work.

This increment changes **two files, +334/−20**: the registry implementation and a self-contained SDK/registry regression file. It imports no Files client, parser, Save implementation or Files-only fixtures. The overall runtime-relative PR is now **83 paths**, all within Desktop apart from the existing root lockfile.

## What the client preserves

- Hosted mode only when the selected source advertises a compatible canonical contract. Older gateways retain the explicit classic Desktop-driven mode; unsupported canonical operations remain unavailable.
- Authoritative history and unfinished state after reopen, without duplicating accepted messages or starting a second local hosted driver.
- Source-qualified participants and speakers across gateways, duplicate profile names and renames.
- Durable Send, Stop, rename, disband, Retry and approval identities, plus ownership checks around setup, grant reconciliation and late creation responses.
- The runtime's extracted round/member/prompt owners and existing classic serial behavior, rather than restoring obsolete inline implementations.

## Verification

Current head: [`68f92bbf8e4b`](https://github.com/dokterdok/hermes-agent/commit/68f92bbf8e4b3cc6b9eff5f5dff2c28704958b48). The previous published head `5e928f766bf2` remains its parent: this is a normal forward update, not a history replacement.

For this two-file correction:

- **21 regression failures and five passing controls** on the previous published production code; all **26 cases pass** with the correction.
- **146 tests pass across nine gateway/SDK files**, with two workers and zero retries. A separate parent run passed the 26 decisive cases; these counts overlap.
- All three no-emit TypeScript projects pass. Scoped ESLint passes with **zero errors and zero warnings**. One renderer-only build passes.
- Production bytes match the independently reviewed correction; standalone tests use the actual SDK-to-registry route, with synthetic native descriptors and socket replies. No upper Files dependency is needed to run them.

This is not a full Desktop suite, live WebSocket, installed Mac/Windows, native Save or two-client acceptance claim. Hosted CI is checked separately against the exact head; absent or approval-required runs are not passes.

<details>
<summary>Earlier forward-integration evidence</summary>

The `5e928f766bf2` forward merge preserved both published histories and reconciled only the Desktop overlaps against the runtime. Its 167 focused cases across eight files, three type projects, scoped lint and renderer build passed. Review also corrected stale adoption after asynchronous hosted creation. Those results belong to that earlier checkpoint and are not added to the current transport total. Original contributor commits and authored history remain ancestors.

</details>

## Dependencies and remaining gates

- **#106742** at `b36398f71929d6069905ccca59af8d1516c5ee3a`: the tested gateway-runtime base, still unmerged.
- **#100016** at `f373dc2157c6e434a27c57cc8996c51b8608aa45`: separately delivered route/target and input/export-consent contracts, with its own declared prerequisites. Root/default source work does not establish named-target or matching-revision native acceptance.
- **#99107 and #96162** remain separate owners. This Desktop PR does not absorb passive/quarantine storage, canonical Stop or cross-owner safety work.
- **#98072, #99159, #104198 and #104199** own Files and Bot output; **#98073** owns messaging controls; **#104601/#105079** own passive preservation and gated recovery.
- Packaged reconnect UAT, exact-head hosted CI, maintainer review and matching-revision integration remain open. Losing Desktop is not losing the authority host: this PR enables neither host takeover nor credential/workspace migration.

#98307 remains permanently draft for composed integration/UAT, not a landing PR. #97681 tracks the remaining cross-layer work. Deliver through the owning PRs rather than copying their implementations into this one.


andrexibiza (commented):
Reviewed the Desktop continuity delta at exact head `8f8c2bdf870a3b2790a8becd013df0aefa624e95`, after #97797 head `1c6bf4cc1a5294d7387bb5dd9eb020dedcef1b44`. Exact-head CI, Docker, and Nix are green.

The setup cleanup journal, persistent command outbox, monotonic replay, mutation fences, partial multi-gateway rollback, and projected-room read-only boundary are coherent. One durable-retry defect remains.

### P1 reliability — transient worker unavailability permanently terminalizes queued commands

`apps/desktop/src/plugins/hermes-bots/hosted-room-runtime.ts::terminalCommandFailure()` classifies every integer RPC code from `4100` through `4199` as terminal except `4115`.

`tui_gateway/methods_groups.py` uses `4123` for **“Group Chat worker is unavailable. Restart the Hermes gateway and try again.”**, including `groups.send` and `groups.disband`. That is a transient lifecycle state: the gateway may be connected while its process-owned hosted-room worker is restarting or stopping.

Because `4123` falls inside the terminal range, a persisted send or disband attempted during that window transitions to `terminal-failure` and is removed from replay rather than retaining the same durable command ID for retry. The outbox therefore converts a recoverable worker restart into permanent command loss.

Required repair:

- replace the broad numeric-range rule with an explicit typed/allowlisted classification;
- make `4123` retryable while preserving the existing command ID and per-room ordering;
- add cold-restart tests in which a persisted send and disband each receive `4123`, survive storage reload, then settle exactly once after the worker returns;
- retain terminal treatment only for errors that prove the command can never become valid without user mutation.

### Additional landing blockers

- This layer inherits the #97712 UTF-8 replay-bound defect and #97797 silent policy/capability auto-reauthorization defect.
- The delta modifies `gateway/hosted_rooms.py` after it exceeds the repository's hard 2,000-physical-line ceiling.

This head is not merge-ready until the retry classification, upstream source blockers, and touched-godfile violation are repaired and re-proven on the recomposed exact head.

dokterdok:
@teknium1 updated #97846 at `c0f6d36a98` so the salvage target is no longer the conflict-heavy pre-merge stack.

This head is directly on `main@1cf3639813` after #99244, preserves jugol's source-authored cross-source creation commit, and removes the `gateway/hosted_rooms.py` / `methods_groups.py` delta that triggered the >2K-owner blocker. The `4123` worker-unavailable path now uses the same durable command ID across reload and retries; terminal failures are explicitly allowlisted. It also keeps remotely deleted Group Chats read-only before local transcript paint and suppresses duplicate Stop while one is pending.

Local proof on the published object: Desktop Hermes Bots 621/621, renderer/Electron/E2E typechecks, and diff checks pass. CI/Docker/Nix are now running; broader packaged UAT can follow without holding back the code salvage.

andrexibiza (commented):
Re-reviewed the recomposed Layer-4 carrier at exact head `c0f6d36a98013911be6b3ebd9c55ac8d2c7e976c` on current `main@1cf36398135f4848a1d04b2167ffb564b7881d35` after #99244 landed.

The prior P1 I raised is repaired in source on this object. `terminalCommandFailure()` now uses the explicit terminal allowlist `{4110, 4111, 4113, 4117}` instead of treating the whole 4100–4199 range as terminal, so worker-unavailable `4123` stays retryable. The exact regression exercises both persisted `send` and `disband`: each receives `4123`, remains pending across runtime/storage reload, reuses the same idempotency key, and settles exactly once after the worker returns. The current tree also carries the terminal-event-before-disband replay and fresh-ID-after-hosted-fallback repairs described by the Layer-4 contract.

The current HEAD itself is hosted-green: CI `33373088357`, Docker `33373087093`, and Nix `33373087065` are all successful. The runtime and its focused test owner are both below the 2K ceiling (`hosted-room-runtime.ts` is 1,708 lines in this diff; `hosted-room-runtime.test.ts` is 1,685).

This six-commit release train is nevertheless **not acceptable yet** under the repository's every-surviving-commit proof gate. I queried hosted workflow evidence for every exact commit in order:

1. `71c42699c308127f3672dc5712ea4e9a0615c9d0` — no CI/Docker/Nix workflow run attached.
2. `ae9ff695287e4246e79b484d3059aa1ff612896e` — no hosted workflow run attached.
3. `dc310b483bdb141f76e15ed1dba2607376c25d64` — no hosted workflow run attached.
4. `9ee68936171a3e1732458d37575576aaf4e8eb73` — no hosted workflow run attached.
5. `433e19551a1fc3409ef0908f88f0218f11af6976` — no hosted workflow run attached.
6. `c0f6d36a98013911be6b3ebd9c55ac8d2c7e976c` — CI/Docker/Nix green as above.

HEAD-green therefore cannot certify this train. Before landing, every surviving commit needs exact-object hosted proof; if the train is recomposed/squashed to obtain that proof, the resulting new objects must be revalidated rather than inheriting this HEAD receipt.

There is also a live merge-order dependency now: #99107 has been rebased directly on #99244/main and is the active passive-replica/takeover safety follow-up. Its security behavior is repaired, but its moved head currently fails the hard modified-file 2K gate in `gateway/hosted_rooms.py`. Layer 4 should consume the final accepted safety object (or an equivalent landed safety extraction), not bypass it. Any restack after that object creates new Layer-4 commit identities and requires fresh every-commit proof.

So the earlier `4123` durability blocker is closed; the remaining acceptance blockers are exact release-train proof and the unresolved upstream safety/2K landing edge.

dokterdok:
@andrexibiza @teknium1 I am closing the every-surviving-commit proof gate without preserving six review objects.

Current head `22b26d04ca` is intentionally the first coherent stage: fully green safety prerequisite #99107 `2558fb7904`, then jugol's source-authored cross-source creation commit. Its focused tests pass 26/26 plus renderer/Electron/E2E typechecks. CI/Docker/Nix are running on that exact object now.

As soon as it is green, I will add one consolidated Desktop continuity commit containing the former five David commits, rerun the full 621-test Bot Mode suite and typechecks, and obtain CI/Docker/Nix on that exact second object. The PR body now labels this temporary proof stage so it cannot be mistaken for the complete feature.

dokterdok:
@andrexibiza @teknium1 the every-surviving-commit proof gate is now closed.

- #99107 safety `2558fb7904`: CI/Docker/Nix green on its own PR.
- jugol source-authored cross-source creation `22b26d04ca`: CI `33381788568`, Docker `33381787708`, Nix `33381787705`, all green.
- consolidated Desktop continuity `52f80019d2`: CI `33382258610`, Docker `33382257942`, Nix `33382257989`, all green.

The final train therefore has two Layer-4 commits rather than six, preserves jugol's Git authorship, keeps every modified owner below 2K, and passes the complete Desktop Bots suite 621/621 plus renderer/Electron/E2E typechecks. The prior `4123` durability blocker remains closed.

andrexibiza (commented):
Re-reviewed the recomposed Layer-4 carrier at exact head `52f80019d24e97e53db4e20578bcd1a417393510` after the prior six-commit proof failure and the #99107 touched-godfile blocker.

Both outstanding acceptance blockers are closed in the current release train.

The branch now has exactly three surviving commits, and each exact object has its own hosted CI/Docker/Nix proof:

1. `2558fb790411293636f324a22775ca485effaaa0` — passive-replica/takeover safety from #99107: CI `33378857063`, Docker `33378856120`, Nix `33378856056` — all successful.
2. `22b26d04ca6719a4e87326681079dc1c040eabbc` — jugol's cross-source creation commit: CI `33381788568`, Docker `33381787708`, Nix `33381787705` — all successful.
3. `52f80019d24e97e53db4e20578bcd1a417393510` — consolidated Desktop continuity: CI `33382258610`, Docker `33382257942`, Nix `33382257989` — all successful.

That replaces the former six-object train where five surviving commits had no hosted proof. The first object is also the current sharded #99107 safety carrier: the former `gateway/hosted_rooms.py` >2K blocker is gone while the reservation/quarantine transaction fence and passive-authority semantics remain intact.

I rechecked the prior Layer-4 P1 as well. `terminalCommandFailure()` still uses the explicit terminal allowlist `{4110, 4111, 4113, 4117}`, so worker-unavailable `4123` remains retryable. The current regression still drives both persisted `send` and `disband` through `4123`, storage/runtime reload, same-command replay, and exactly-once settlement after recovery. The deleted-room read-only fence, duplicate-Stop suppression, terminal-event/disband ordering, and fresh-ID fallback protections remain present in the current object.

Live upstream `main` is now `446563262e54f03b5632f5590ee5dbdad143739c`. The two commits added after this train's `3aee290899` foundation touch the Desktop registry-primary profile-scope path (`apps/desktop/electron/main.ts` and `registry-primary-profile-scope.test.ts`); those paths are disjoint from #97846's changed-file set, and GitHub currently reports #97846 mergeable.

I do not find a remaining P1-P3 code, topology, every-commit proof, or structural blocker on this exact three-object carrier. If any object is rewritten again before landing, the rewritten train needs fresh exact-object CI/Docker/Nix rather than inheriting these receipts.

andrexibiza (commented):
Landing-edge correction to my immediately preceding exact-head review: upstream `main` moved during the review window and is now `38b7d0f4cf8a11137d8da5e3d95d7b5b7e41fb46`, six commits beyond the `446563262e...` snapshot cited there.

I compared that exact interval. The new main work is credential/env/config/web-state repair plus contributor mappings; it introduces no hosted-room gateway path, no Desktop Hermes Bots path, and no `contributors/emails/solomoj94@gmail.com` collision. #97846 remains open at exact head `52f80019d2`, still contains the same three surviving commits, and GitHub still reports it mergeable.

So the earlier main-SHA sentence is stale, but the release-train verdict is unchanged: all three surviving objects still have their own successful CI/Docker/Nix receipts, the #99107 safety object is unchanged, and the Layer-4 `4123` durability repair remains on the reviewed head. Any subsequent head rewrite still requires fresh exact-object proof.

andrexibiza (commented):
Re-reviewed the moved release train at exact head `8b59f5bf3f8fb0c1c02947f79a3f4972f3c29f90` against current `main@04fea67963fa36798fe215b9d3c7b5d787a632b4`.

The earlier every-surviving-commit proof blocker is closed: the train is now exactly `2558fb7904 -> 22b26d04ca -> 52f80019d2 -> 8b59f5bf3f`, and all four exact objects have successful CI/Docker/Nix runs. The new head is green in CI `33420171407`, Docker `33420170418`, and Nix `33420170581`. The 207-commit main-only interval is path-disjoint from this PR's changed-file set, GitHub reports the PR mergeable, and there are no unresolved formal review threads.

The reconnect repair itself closes the stale-Ready-state UX, but it introduces one P1 durability/security hole in the fresh-grant lifecycle.

### P1 — reconnect mints an unjournaled room grant before all failure/crash points

`reconnectHostedGroupChatPeer()` calls `groups.peer.invite` first. That RPC immediately mints the grant **and** persists its scope with `reserve_peer_room()`. The reconnect path then performs client-side catalog/target/profile/URL validation and only installs revocation inside the later `groups.peer.register` catch.

That leaves multiple post-effect exits with a live grant and no durable cleanup owner:

- the process can die after `groups.peer.invite` returns and before `groups.peer.register`;
- catalog / installation / digest / target-URL validation can reject after the grant was minted;
- `groups.peer.revoke` can itself fail during the registration catch, and that failure is currently swallowed.

This is exactly the lifecycle class the existing hosted-room setup path already handles with the persisted `HOSTED_ROOM_CLEANUP_KEY`: after a peer invite, it records a `peer-revoke` cleanup operation before continuing, then releases/arms that journal according to settlement. Reconnect bypasses that mechanism entirely. A crash or transient revoke failure can therefore leave the newly minted room/profile capability valid until TTL expiry with no restart path that retires it.

Required repair: bind reconnect to the same durable peer-revoke journal (or an equivalent persisted cleanup owner) **immediately after invite and before any later validation/registration step**; retire/release it only after the home has durably adopted the new route; on any failure, leave cleanup armed for restart rather than best-effort swallowing revoke failure. Add regressions for (1) validation failure after invite, (2) revoke failure, and (3) storage/runtime restart with the cleanup operation pending, proving the grant is eventually revoked and never silently orphaned.

The reconnect behavior and exact-object CI are otherwise coherent, but this capability-lifecycle gap keeps the new head from the acceptance gate.

andrexibiza (commented):
Landing-edge correction to the exact-head review above: `main` advanced again during this sweep and is now `c74cf2333c3b6430896dfb0a5b9e42dd96bf0dfe`, 10 commits beyond the `04fea67963...` snapshot I cited.

I compared that exact 10-commit interval. It touches compression/deadline/scheduler/gateway-status/update/process/WhatsApp and adjacent tests, but none of #97846's changed Group Chat / hosted-room paths. The PR head is still exactly `8b59f5bf3f8fb0c1c02947f79a3f4972f3c29f90`, GitHub still reports it open and mergeable, and the every-surviving-commit CI/Docker/Nix receipts remain attached to the same four objects.

So only the main SHA/path-drift sentence changes. The substantive verdict does not: the new reconnect path still needs durable post-invite `peer-revoke` cleanup ownership before the fresh grant can be considered safely settled.

andrexibiza (commented):
Final landing-edge correction after another `main` move: current `main` is `b8d3c8aa07301eeacad06cde25e460a189c9d6d6`, six commits beyond `c74cf2333c`.

I compared the exact `c74cf2333c..b8d3c8aa07` interval and the complete 223-commit main-only FILE-LIST from this PR's merge base. There is no exact path collision with #97846's 31 changed files. The head remains exactly `8b59f5bf3f8fb0c1c02947f79a3f4972f3c29f90`, and its four surviving objects retain their individual green CI/Docker/Nix receipts.

One source-of-truth wrinkle matters for landing claims: the current normalized PR metadata now reports `mergeable=false`, while GitHub's raw PR payload is still lagging on an older base snapshot and reports `mergeable_state=clean`. I am therefore withdrawing mergeability as a current acceptance receipt rather than choosing the favorable cached answer. A current-main restack/recalculation is required before landing.

The substantive code blocker is unchanged and independent of that metadata race: reconnect still mints/persists the fresh scoped grant before durable `peer-revoke` cleanup ownership exists. The grant-lifecycle repair and its restart/revoke-failure regressions remain required before this train is acceptable.

dokterdok:
Addressed the reconnect capability-lifecycle review in consolidated head `50d424e37a` (one commit replacing the earlier reconnect cut).

- cleanup is read-back verified before any post-invite work and survives Desktop/plugin restart;
- reconnect uses scope-plus-grant exact revocation, so a losing Desktop cannot invalidate the winning route;
- response-loss reconciliation requires the persisted route's matching grant SHA-256; `unavailable`/unknown remain pending;
- stale success and rejection paths are lifecycle/poll-generation fenced;
- older gateways do not expose Reconnect until they advertise the exact-revoke contract.

Validation: 639 Desktop Bot Mode tests, all Desktop typechecks/lint, 77 backend service/RPC/storage tests, Ruff/diff checks. Two independent Codex subagent reviews now report SHIP with no P1/P2. Fresh exact-head CI/Docker/Nix is running.

andrexibiza (commented):
Re-reviewed the moved head at `50d424e37a055d0b35784a0eb4933aae46e38e77` against the prior fresh-grant cleanup blocker.

The blocker is resolved. The reconnect path now journals the exact fresh grant as `peer-revoke-exact` immediately after `groups.peer.invite`, before hashing/catalog/profile/URL validation; a journal-persist failure exact-revokes immediately, validation failure arms + dispatches the durable cleanup, and the persisted cleanup path survives Desktop restart and retries failed exact revocation. The regression set covers invalid invitation after grant creation, restart replay, lost registration response, concurrent winner reconciliation, definitive rejection + failed revoke retry, and preserving an ambiguous/unavailable committed route rather than revoking it.

Release proof is also clean for every surviving commit, not only HEAD: `2558fb7904` CI/Docker/Nix `33378857063 / 33378856120 / 33378856056`; `22b26d04ca` `33381788568 / 33381787708 / 33381787705`; `52f80019d2` `33382258610 / 33382257942 / 33382257989`; `50d424e37a` `33432814495 / 33432813771 / 33432813795` — all successful. There are zero unresolved inline review threads.

I also checked the 32-commit `main` drift from `5505042f40` through current `8dbf07e950`; none of this PR's Group Chat continuity file families appear in that interval. GitHub currently reports the PR mergeable.

No remaining P1/P2 finding on this exact object.

andrexibiza (commented):
Landing-edge correction: `main` advanced again after the preceding review and is now `d95682cd04ab16314fb44f9ad360144806b64bed`, three commits beyond the `8dbf07e950` snapshot cited there.

I compared the exact `8dbf07e950..d95682cd04` interval. It changes session-listing/resume surfaces, locales, docs, and their tests; it does not touch this PR's Desktop Hermes Bots, hosted-room, or Group Chat continuity file families. The PR head remains exactly `50d424e37a055d0b35784a0eb4933aae46e38e77`, GitHub still reports it mergeable, and the four surviving exact objects retain their successful CI/Docker/Nix receipts.

The substantive verdict is unchanged: the fresh-grant cleanup blocker is repaired on this exact head, and I find no remaining P1/P2 finding.

dokterdok:
Exact head `13115dd95f` closes the final reconnect concurrency findings:

- a live Desktop owns its compensation journal through a per-window Web Lock; another window can claim it only after stop/crash, with a bounded lease fallback for older runtimes;
- the home gateway compare-and-swaps the observed grant fingerprint under the same lock as persistence/publication, so stale recovery cannot overwrite a newer winner or a concurrent refresh;
- route status and fingerprint are one atomic snapshot;
- a late invitation response remains durably journaled after runtime stop;
- a lost registration response remains an explicit unknown outcome and is never reported as success.

Validation on this exact object: 66 Bot Mode files / 642 tests, 98 focused backend service/RPC/storage tests, all Desktop typechecks and lint, Ruff and diff checks. Two independent Codex adversarial reviews report SHIP with no P0/P1/P2. Required CI, Docker amd64/arm64, and Nix are all green.

andrexibiza (commented):
Exact-head re-review on `13115dd95f92a9ade9d88ce7c1c03b704187fa30` after the reconnect commit was replaced.

The surviving train is hosted-green commit-by-commit: `2558fb7904` (CI 33378857063 / Docker 33378856120 / Nix 33378856056), `22b26d04ca` (33381788568 / 33381787708 / 33381787705), `52f80019d2` (33382258610 / 33382257942 / 33382257989), and `13115dd95f` (CI 33443646095 / Docker 33443645339 / Nix 33443645330). I also re-read the replacement reconnect path, durable cleanup journal, grant-fingerprint CAS, exact-grant revocation, and the new regressions. The replacement modules stay under the 2K ceiling; there are no inline review threads. Current landing edge is `main@180291162ff4df0d42b5dc4fecd08005cf7cebf9`; the base→main drift contains no `hermes-bots` or `hosted_room` paths, so this moved-head reconnect slice has no current-main FILE-LIST collision.

One substantive defect remains:

**P2 — reconnect lookup drops the source-qualified identity and can select the wrong gateway.** `matchingLocalMember()` in `hosted-room-reauthorization.ts` matches only `(profile, handle)` and returns the first local member. Cross-source Group Chats explicitly allow the same profile/handle to exist on different connections; the room keeps those members source-qualified. If two peer gateways both contribute (for example) `builder`/`builder`, reconnecting the second server member can pick the first local descriptor, probe that first gateway, then fail the later `peerCapability.authorityId !== targetAuthority` check. This fails closed, but the affected Bot can never be reauthorized even though the server member already carries the exact target installation ID. Resolve the local member by the server target authority/connection as part of the match (not only profile+handle), and add a regression with duplicate profile+handle on two peer gateways that reconnects the second member.

The PR body also correctly leaves the **new packaged reconnect path** as the final exact-head UAT gate; the existing field UAT was on `15afb8fb01` and predates this replacement reconnect implementation. I would not treat that older UAT as acceptance proof for `13115dd95f`.

dokterdok:
Addressed the duplicate-source reconnect finding at `a0a032bfc5`. Reconnect now resolves matching profile/handle candidates by the exact target installation authority; ambiguous candidates fail before capability probing, and a stale cached authority is revalidated live before any invitation, grant, registration, or success refresh. The three source-qualification regressions pass alongside the complete Desktop Bot Mode suite (**66 files / 642 tests**), renderer/Electron/E2E typecheck, and diff check. Two independent exact-diff re-reviews report **SHIP** with no P0/P1/P2. The packaged reconnect UAT remains explicitly pending.

andrexibiza (commented):
Re-reviewed exact head `e75e10c62a6319e907c77f4a1fcd415648292ce3`, including the new source-qualification repair `a0a032bfc52643b20bb4f9a3123738c1c7d3ca94`, against final-checked `main@5fae0d243f98a81e49663d4c48b2ed871b9a14c2`.

The source P2 from my prior review is closed. `matchingLocalMember()` no longer accepts the first `(profile, handle)` match: it qualifies duplicate candidates by the server member's exact target installation authority, returns a sole candidate only when unambiguous, and otherwise fails closed. The new regressions cover the positive two-gateway duplicate-profile case, missing authority qualification, and stale cached authority; the stale path live-revalidates capabilities before invitation/grant/registration. I do not see a new P0/P1/P2 in that delta.

The release train is still not acceptable under the every-surviving-commit proof gate. The six submitted commits are `2558fb7904`, `22b26d04ca`, `52f80019d2`, `13115dd95f`, `a0a032bfc5`, and merge head `e75e10c62a`. The first four already have exact-object CI/Docker/Nix success receipts. The new repair commit `a0a032bfc52643b20bb4f9a3123738c1c7d3ca94` has Docker `33589251600` success and Nix `33589251682` success, but required CI `33589250712` is `failure` and dispatched **zero jobs**. The later merge head is green — CI `33590874227`, Docker `33590873674`, Nix `33590873776` — but HEAD-green cannot certify a surviving red commit.

Required release repair: recompose/squash so `a0a032bfc5` is no longer a surviving submitted object (while preserving source authorship/credit), then require fresh CI/Docker/Nix success on every surviving replacement commit. Do not inherit proof from `e75e10c62a` onto new commit identities.

The PR's own packaged reconnect UAT gate also remains explicitly outstanding; the older `15afb8fb01` field receipt predates this replacement reconnect path and cannot certify it.

Landing-edge cross-check: `main` is four commits ahead of the branch's merged `ff7745fb0a` base. Those four commits touch only `agent/conversation_loop.py`, `tests/run_agent/test_413_compression.py`, and `website/docs/reference/faq.md`, disjoint from this PR's FILE-LIST, and GitHub currently reports the PR mergeable. No inline review threads exist to resolve.

So: the duplicate-source reconnect defect is substantively fixed; the remaining hard acceptance defect is exact-object release proof, plus the PR's declared packaged reconnect UAT.

andrexibiza (commented):
Landing-edge correction to my immediately prior review only: `main` advanced again while the review was being settled, to `273461f1393f0be5ecd08415779804c755c731f6`. Relative to #97846's merged `ff7745fb0a82be914f490f14967e158347361dd6` base, `main` is now eight commits ahead and touches `agent/conversation_loop.py`, `gateway/run.py`, `gateway/scale_to_zero.py`, `tests/gateway/test_scale_to_zero_dashboard_client.py`, `tests/run_agent/test_413_compression.py`, `tui_gateway/ws.py`, and `website/docs/reference/faq.md`. That FILE-LIST remains disjoint from this PR's 40 changed files, and GitHub still reports #97846 mergeable.

The substantive conclusion is unchanged: the duplicate-source reconnect P2 is fixed, but surviving commit `a0a032bfc5` still has failed required CI `33589250712` with zero dispatched jobs, so the submitted train is not every-commit green; packaged reconnect UAT also remains pending.

andrexibiza (commented):
Exact-head re-review on `240f42ff4f54428a7a4b59309d56ebd0312d42d5`.

The previous source-owner qualification defect is closed on this object. `matchingLocalMember()` now binds each duplicate alias candidate to its cached gateway source, queries that gateway's authoritative room-member record, and accepts only the candidate whose `targetInstallationId` matches the current local registered installation; stale, no-exact, and multiple-exact candidates fail closed. The two-gateway regression with identical profile + handle selects only the single source whose remote row matches the current installation, with additional stale-cache and ambiguous-source regressions.

Every surviving commit is exact-SHA green:

- `2558fb7904` — CI `33378857063`, Docker `33378856120`, Nix `33378856056`
- `22b26d04ca` — CI `33381788568`, Docker `33381787708`, Nix `33381787705`
- `52f80019d2` — CI `33382258610`, Docker `33382257942`, Nix `33382257989`
- `13115dd95f` — CI `33443646095`, Docker `33443645339`, Nix `33443645330`
- `240f42ff4f` — CI `33602385323`, Docker `33602383494`, Nix `33602383484`

Landing-edge cross-check: live `main@bfbb34bbec25d40f4982a70748ba390ff5efb363` is 17 commits beyond submitted base `fdb2e10a8e606558e10ee5305ca7b74fbb2a3f24`; GitHub currently reports `mergeable=true`, and the current main-only FILE-LIST does not intersect this PR's touched paths.

No new P0/P1/P2 found in this repair. The source/proof train is green.

I am not transferring the older packaged-field UAT to this exact head. The PR's own acceptance record still leaves packaged reconnect UAT for the new path as the final exact-head acceptance gate.

teknium1:
Thanks for the substantial work here, and for keeping the local serial round engine intact.

We're not ready to take group-chat continuity yet. Group chat on main is still settling (we reverted parallel rounds to serial this week and are working through the open Bot Mode bugs), and landing a hosted-room client of this size on top would complicate that before the base is stable. Leaving this open rather than closing it; we'll revisit in about a month once the current Bot Mode / group-chat bug cluster is cleared.

Two things that would make that review tractable when we come back: keep #99107 and #96162 as their own PRs rather than carried in this train, and split the `hosted_rooms.py` -> `hosted_room_storage.py` relocation into a declared refactor PR (or drop it) so the feature diff is just the Desktop client.

dokterdok:
> Thanks for the substantial work here, and for keeping the local serial round engine intact.
> 
> We're not ready to take group-chat continuity yet. Group chat on main is still settling (we reverted parallel rounds to serial this week and are working through the open Bot Mode bugs), and landing a hosted-room client of this size on top would complicate that before the base is stable. Leaving this open rather than closing it; we'll revisit in about a month once the current Bot Mode / group-chat bug cluster is cleared.
> 
> Two things that would make that review tractable when we come back: keep #99107 and #96162 as their own PRs rather than carried in this train, and split the `hosted_rooms.py` -> `hosted_room_storage.py` relocation into a declared refactor PR (or drop it) so the feature diff is just the Desktop client.

Thanks, that makes sense. I’ll keep #99107 and #96162 separate and recut #97846 on current main so it contains only the Desktop continuity client, with no storage relocation. I’ll keep the field stack available for end-to-end testing and revisit continuity in about a month, once the current Group Chat bugs have settled. In the meantime, I’ll see where I can help with that Bot Mode bug cluster.

dokterdok:
## Owner-only pair and queue-ownership delivery

The current #97681 / #106742 dependent proposals are now publicly reproducible:

- **#104198:** unchanged accepted head `3cc3eb71b011c1d58aa6e87e1bda3e422093a074`, three owner commits / 14 paths, default/root classic scope. [Backend composition/build recipe](https://github.com/dokterdok/hermes-agent/blob/3cc3eb71b011c1d58aa6e87e1bda3e422093a074/docs/development/classic-owner-composition.md).
- **#97846:** normal-forward update `967cc64e306af0c555d873f071851476c1284808`, including immutable cancellation recipe pin `f3c672821059d7ae37643dbe96b520c824091bab`. The owner-distinct repair and lower-first composition have independent source acceptance. Command/fence ownership is distinct from thread coalescing and active backend-session ownership; recovery, rename, exact Stop/lease loss, both same-thread enqueue orders and unrelated-work isolation are covered by existing causal evidence. [Lower-owner contract and correction](https://github.com/dokterdok/hermes-agent/blob/967cc64e306af0c555d873f071851476c1284808/apps/desktop/scripts/delivery/97846/README.md).
- **#104199:** existing PR reconstructed at `dad51cfe4adfd23c71b4d6d3c62b6991f94552dd` and retargeted to #106742 pin `485d5f6848c2598ca00fa7704e50e8ae66d0983a`. Its entire public interval is 11 owner commits / 35 paths; every public commit and changed-file blob was checked. No lower implementation commits are bundled. The [original head and contributor history](https://github.com/dokterdok/hermes-agent/commit/03606761abc01ada4b15c7f9632d27a4e6011b41) remain reachable through the verified backup branch; replacement used the exact expected-old-head lease.

### Executed public recipe

[Desktop recipe](https://github.com/dokterdok/hermes-agent/blob/dad51cfe4adfd23c71b4d6d3c62b6991f94552dd/apps/desktop/scripts/delivery/104199/compose.sh) · [exact pins, order, setup/build commands and limits](https://github.com/dokterdok/hermes-agent/blob/dad51cfe4adfd23c71b4d6d3c62b6991f94552dd/apps/desktop/scripts/delivery/104199/README.md).

After publication, this exact script ran in `compose` mode with all six remote/ref overrides removed. It fetched only the declared public GitHub refs, checked the immutable pins and intermediate stages, and exited 0 with final tree `f86ea20723955476e1d94297fe0a334ab9199dba`. Lower cancellation was applied before Files. The result is identical to the independently accepted composition already exercised with 21 selected causal passes, three TypeScript projects, scoped lint, renderer/Electron build and asset assertion. Lower-only and composed test selections overlap, not additive unique coverage. Public reconstruction did not replay those unchanged suites or claim a second build/test run.

These are dependent owner proposals, not standalone builds; unmerged prerequisites do not require a new PR or combined upstream branch. The inherited availability fixture failure remains disclosed in the recipe/PR. No new native Save/Cancel, installed-client, whole-suite, live end-to-end or exact-head hosted-CI success is claimed. The complete seven-layer programme is still open; this closes the bounded queue-ownership and thin-pair publication milestone, not the whole programme.

apps/desktop/src/plugins/hermes-bots/shipped-group-adoption.ts
```diff
@@ -0,0 +1,1435 @@
...
+    const imported: ShippedGroupImportMember = {
+      source_member_id: sourceIdForMember,
+      name,
+      profile,
+      handle: uniqueHandle(text(member.handle) || profile, sourceIdForMember, usedHandles),
+      // Released persistence intentionally labels every member remote-capable.
+      // Backend import needs a different fact: remote relative to the selected
+      // canonical owner connection.
+      remote_source: Boolean(connectionId && connectionId !== ownerConnectionId),
+      active: true,
```
andrexibiza: **P1 — preserve missing source ownership instead of importing an active local replacement.** An orphan with `remoteSource: true`, `sourceMissing: true`, and no retained connection ID reaches this code as `connectionId === ''`; these lines export it as `remote_source: false, active: true`. `hygiene.ts::annotateOrphanedGroupChatMembers` explicitly recognizes that persisted shape, and `durableGroupChatMembers` preserves its missing-owner mark. Another intact member can still select the room owner, so the room need not be owner-ambiguous. The new import payload then loses the distinction between that orphan and a genuine owner-local legacy profile. Keep unresolved/scoped-missing ownership held and non-executable (or refuse adoption without changing the source), rather than allowing a same-named local profile to become the replacement. My extracted mapper probe reproduced this with a valid owner member alongside the orphan; normal known-local/known-other-source controls pass. Add the hydrate/annotation/import composition regression and prove zero ready/admission transition for the orphan.

andrexibiza (commented):
## Exact-head review: two P1 owner-identity blockers

Reviewed `c1cf8225f9d7c28d22aa204fe20f7c399d4700da` against the PR's reported runtime base `b36398f71929d6069905ccca59af8d1516c5ee3a`. Main was observed at `4dc7a23690bbf242aaa6ecb8351abce97cb0ad3d`; that is a separate observation, not a claim that this stacked branch has been accepted on current main. The PR is still open/draft and GitHub currently reports it non-mergeable.

The adopted-route lease and the decision to retain uncertain work rather than replay it are the right foundations. However, owner qualification is still lost at two boundaries in the new adoption work. These are source defects, separate from the already-disclosed Files/composition fixture failures.

### P1 — a source-missing member is imported as active and owner-local

In [`buildShippedGroupImport`](https://github.com/dokterdok/hermes-agent/blob/c1cf8225f9d7c28d22aa204fe20f7c399d4700da/apps/desktop/src/plugins/hermes-bots/shipped-group-adoption.ts#L384-L410), an empty `memberConnectionId(member)` produces `remote_source: false`, and every retained roster member receives `active: true`. Neither `sourceMissing` nor the unresolved source-qualified state prevents that conversion.

This is an existing persisted shape, not a malformed-input invention. [`annotateOrphanedGroupChatMembers`](https://github.com/dokterdok/hermes-agent/blob/c1cf8225f9d7c28d22aa204fe20f7c399d4700da/apps/desktop/src/plugins/hermes-bots/hygiene.ts) explicitly handles a remote/scoped descriptor that has lost its connection ID. [`durableGroupChatMembers`](https://github.com/dokterdok/hermes-agent/blob/c1cf8225f9d7c28d22aa204fe20f7c399d4700da/apps/desktop/src/plugins/hermes-bots/group-membership.ts#L422-L474) deliberately preserves its degraded mark and missing route. Startup performs that hygiene before adoption. Another intact member can still provide an unambiguous owner for the room.

With a known `owner-a` member alongside `{ name: 'builder', remoteSource: true, sourceMissing: true }` and no retained connection ID, the mapper exports the orphan as:

```json
{"profile":"builder","remote_source":false,"active":true}
```

There is no `connection_id` or preserved missing-owner classification in that member's import payload. A source-qualified descriptor with its connection missing takes the same path. The importer is therefore told that an unresolved former source is an active local member; it no longer has the evidence needed to distinguish that orphan from a genuinely local legacy profile. A same-named profile on the selected owner must not become the replacement execution target merely because the original connection disappeared.

**Required:** preserve unresolved/missing ownership as held/non-executable, or refuse adoption while keeping the original room intact. Do not solve this by marking every `remoteSource` row foreign: released persistence also marks valid owner-local rows remote-capable. The distinction must be positively established owner equality versus known-other versus unresolved ownership. Add a hydrate → orphan annotation → adoption regression with a same-named local profile, retaining normal owner-local and known-other-source controls. The matching backend composition should prove that the orphan never becomes ready or creates an admission.

### P1 — an ordinary canonical alias bypasses an adopted pending Send's owner checks

[`readCanonicalGroupSend`](https://github.com/dokterdok/hermes-agent/blob/c1cf8225f9d7c28d22aa204fe20f7c399d4700da/apps/desktop/src/plugins/hermes-bots/canonical-group-send.ts#L61-L78) checks the persisted installation/source/hash only when the *incoming binding* has `adoptionOwner`. A pending entry containing those qualifications is nevertheless returned when the incoming binding has none.

That weaker binding has a production producer: [`registerCanonicalGroup`](https://github.com/dokterdok/hermes-agent/blob/c1cf8225f9d7c28d22aa204fe20f7c399d4700da/apps/desktop/src/plugins/hermes-bots/canonical-group-registry.tsx) registers discovered rooms using only connection/profile/room ID, without the adopted binding's owner or lease. The canonical list is exposed by the Bots pane. Both aliases address the same Send journal key, which excludes adoption qualification.

The resulting path is:

1. Prepare a pending Send under the adopted binding for installation A.
2. Open the same room through its ordinary canonical-list alias. It restores that qualified entry despite having no `adoptionOwner` or route lease.
3. Change the saved connection to installation B while that view remains mounted.
4. Retry the restored Send. [`canonicalGroupRequest`](https://github.com/dokterdok/hermes-agent/blob/c1cf8225f9d7c28d22aa204fe20f7c399d4700da/apps/desktop/src/plugins/hermes-bots/canonical-groups.ts#L137-L171) takes the descriptor-only `host.requestProfile` path, and the workspace's optional `isCurrent` guard does not invalidate this binding. A's retained payload can be dispatched to B. B rejecting an unknown room afterward cannot undo disclosure of the request body.

The offline diagnostic reproduced the journal downgrade and a recorded `groups.send` to a simulated replacement installation, with A's original event ID and payload. The controls correctly reject a mismatched owner *when qualification is supplied*, and reject an expired original lease before dispatch. The missing boundary is the weaker alias, not those working checks. The legacy descriptor path predates this increment; the new defect is allowing the newly qualified adoption intent to escape through it. This is also distinct from the earlier shared-remote lookup race: a per-request lookup fence does not pin a previously prepared intent's installation.

**Required:** a qualified journal entry must require matching, independently current owner qualification; absence cannot bypass comparison. Retain the original intent on refusal. Canonical aliases for the same adopted room must reuse or freshly establish the appropriate installation-bound, generation-owned binding before becoming writable. Do not deserialize a journal record into runtime authority. Cover alias reopen, saved-connection replacement, zero cross-owner sends, unchanged retained intent, and successful retry against the original current owner.

### Verification and limits

I checked the complete changed-file inventory and the `967cc64e306a → c1cf8225f9d7` comparison: nine forward commits and 36 changed Desktop-source paths within the 32-commit/100-file PR. Detailed source tracing covered adoption/checkpoints, orphan persistence, canonical registry/workspace/Send, plugin lifecycle, and the lease-versus-descriptor request boundary. This is not a claim that every path in the larger stack has independent acceptance.

The offline diagnostic ran **12 cases: eight passing controls, four failing safety assertions across the two defect classes, zero unexpected failures**. It exits 1 deliberately. The complete Send and request modules were checked against Git blobs `43af37305f8ce03a94e95a1ece376b6114978777` and `54ab2ad466a75a46f80835e11269541f56de9d9f` before execution. The mapper diagnostic uses the unchanged `buildShippedGroupImport` function excerpt with test-supplied helpers and empty history/held-work inputs. Native journal/RPC boundaries are mocked; no backend execution, mounted React acceptance, installed Desktop, or full repository suite is claimed. These counts are independent of, and not added to, the PR author's local test totals.

Exact-head GitHub queries currently return **zero workflow runs, zero check runs, and no commit statuses**. There is no hosted-green receipt for this object. Earlier accepted heads and the September 22 public cancellation/Files recipe do not certify this new adoption composition.

### Ownership and landing edges

Keep #106742 as the runtime/execution owner; this repair does not require a second Desktop executor. #99107 now advertises the newer history-preservation correction at `82bf61fdec2b2d25737eed57cfb16c310eabca8b`, so the PR body's older retention edge needs matching-revision composition rather than inherited acceptance. The Files/classic owners, including #104198/#104199, remain complementary dependencies, not implementations to copy into this branch. The local importer/shared pin and contextual union in this PR are explicitly not a public-only recipe.

Preserve dokterdok's forward history and the separate contributors' ownership. Fix these two Desktop boundaries, then re-prove the exact resulting owner and public composition. Hosted CI, matching Retention/importer compatibility, packaged reconnect, native Save/Cancel, installed Mac/Windows, and two-client gates remain separate and open under #97681. This COMMENT review is not approval or programme closure.

apps/desktop/src/plugins/hermes-bots/canonical-group-send.ts
```diff
@@ -60,6 +67,14 @@ export async function readCanonicalGroupSend(binding: CanonicalGroupBinding): Pr
     throw new Error('Invalid canonical group Send entry')
   }
 
+  if (entry && binding.adoptionOwner && (
+    entry.binding.authorityGatewayId !== binding.adoptionOwner.authorityGatewayId ||
+    entry.binding.sourceId !== binding.adoptionOwner.sourceId ||
+    entry.binding.requestHash !== binding.adoptionOwner.requestHash
+  )) {
```
andrexibiza: **P1 — a weaker binding must not bypass a qualified pending intent.** This condition validates the stored authority/source/hash only when the incoming binding supplies `adoptionOwner`. A Send prepared under an adopted binding is returned unchanged to an address-only binding. `registerCanonicalGroup()` creates exactly that second binding shape for the same discovered room, and both use the same connection/profile/room journal key. After the alias restores A's pending intent, replacing the saved connection lets its lease-free workspace retry through `host.requestProfile` to B. The hash-verified Send/request modules reproduced that dispatch in an offline recording-RPC harness; explicit wrong-owner and expired-lease controls correctly refused. Require live matching qualification whenever the entry is qualified, retain the intent on refusal, and make writable canonical aliases reuse or establish the installation-bound lease. Do not reconstruct authority from the journal itself. Regress alias reopen → connection replacement → retry, asserting zero cross-owner dispatch and preserved original intent.

apps/desktop/src/plugins/hermes-bots/shipped-group-adoption.ts
dokterdok: Fixed in [a0af96739e7e](https://github.com/dokterdok/hermes-agent/commit/a0af96739e7e7e20ae9dd73359a30b5de9fd14fd).

The mapper now refuses adoption before import if any retained member is source-missing or is remote/source-scoped without a retained connection ID. It preserves the original roster/history rather than exporting a local replacement. It still distinguishes positively matched owner-local from known-other members; valid owner-local rows marked remote-capable are not classified as foreign.

New regressions exercise actual cold hydration → orphan annotation → adoption for three released shapes with a same-named owner-local profile. They assert no import and unchanged original roster/history, with valid local/known-other controls. A separate actual-producer export and real backend-dispatcher probe pass on local importer/Retention composition `c7f03671123f24342adc7be8600f72f286ea154f`: the three refusals create no room or task; the valid control imports idempotently with local members ready, the other-source member held and zero tasks. The backend probe uses a synthetic principal and excludes Output; it is not authenticated live transport or public-only/native proof. Its safety/replica modules match Retention `82bf61fdec2b2d25737eed57cfb16c310eabca8b`; the legacy importer is the composed reconciliation, not an identical blob.

Verification at `a0af96739e7e7e20ae9dd73359a30b5de9fd14fd`: the six new safety cases fail on the old production code; the corrected focused suite passes **38 cases / five files**, plus scoped ESLint, renderer no-emit TypeScript and diff checks. One independent source-only review found no concrete reachable P1/P2 issue; it ran no tests. No implementation was delegated.

Scope/limits: this is a normal forward correction preserving the prior history. The PR remains draft with its pre-existing runtime-base conflict. Exact-head hosted queries currently show zero workflow runs/check runs/commit statuses, not a hosted pass. Public-only composition, full cross-layer acceptance, packaged/native acceptance and multi-window races remain open. No merge or deployment.

apps/desktop/src/plugins/hermes-bots/canonical-group-send.ts
dokterdok: Fixed in [a0af96739e7e](https://github.com/dokterdok/hermes-agent/commit/a0af96739e7e7e20ae9dd73359a30b5de9fd14fd).

The journal now detects qualification from the retained entry itself: absence of incoming qualification cannot bypass comparison. Reading, preparation and retirement require matching independently current adoption identity and route lease, and refusal leaves the original intent unchanged. Canonical discovery returns the adopted room's retained key rather than creating a weaker alias. When no lease exists that key opens the existing read-only recovery view; pre-adoption aliases are invalidated. The request boundary independently refuses an adopted descriptor without a live lease. Neither the journal nor a checkpoint constructs runtime authority.

Mounted workspace regressions reopen via discovery and cover replacement installation → Retry with zero cross-owner sends and an unchanged retained event/payload, plus successful retry under the original current owner. They also cover revocation/read-only fallback and pre-adoption alias invalidation. A native atomic-file journal regression rejects address-only, owner-without-lease and expired-owner access without overwriting or retiring the retained intent. RPC/descriptor boundaries in these client tests are synthetic; installed native/multi-client acceptance is not claimed.

Verification at `a0af96739e7e7e20ae9dd73359a30b5de9fd14fd`: the six new safety cases fail on the old production code; the corrected focused suite passes **38 cases / five files**, plus scoped ESLint, renderer no-emit TypeScript and diff checks. One independent source-only review found no concrete reachable P1/P2 issue; it ran no tests. No implementation was delegated.

Scope/limits: this is a normal forward correction preserving the prior history. The PR remains draft with its pre-existing runtime-base conflict. Exact-head hosted queries currently show zero workflow runs/check runs/commit statuses, not a hosted pass. Public-only composition, full cross-layer acceptance, packaged/native acceptance and multi-window races remain open. No merge or deployment.

andrexibiza (commented):
Exact-head re-review of `a0af96739e7e7e20ae9dd73359a30b5de9fd14fd` against the PR's runtime base `b36398f71929d6069905ccca59af8d1516c5ee3a`.

The two prior P1s are closed in their isolated paths on this head:

- `buildShippedGroupImport()` now refuses source-missing / unresolved source-qualified members before import instead of reclassifying them as active owner-local rows.
- The Send journal now treats persisted owner qualification as a constraint, requires an independently current adopted binding + route lease to read/replace/retire it, and the request boundary refuses adopted bindings that have lost that lease. Successful adoption also retires the earlier descriptor-only alias.

There is still one P1 across the composition of those fixes.

### P1 — a held/refused adoption can leave its pre-adoption canonical alias executable

The new alias retirement in `bindAdoptedCanonicalGroup()` only runs after adoption has successfully reached the adopted binding. But the production flow establishes `shippedAdoption` ownership earlier: `processGroup()` persists the waiting checkpoint before owner/build/import completion, and the new source-owner refusal can then exit through `persistIssue()` without ever calling `bindAdoptedCanonicalGroup()`.

That matters because the earlier canonical-list alias has a different key (`canonical:<connection>:<profile>:<roomId>`). `persistIssue()` only calls `revokeCanonicalGroupBinding(group)`, so it removes the retained room key, not that alias. The alias therefore remains in `$canonicalGroupBindings` and its `isCurrent()` remains true.

The other side is still open as well: `registerCanonicalGroup()` only recognizes a retained adoption when `adoption.route` is already present. A waiting/held checkpoint has no route, so discovery can mint the same descriptor-only alias again while `shippedGroupAdoptionOwnsExecution()` says the room is held.

That alias bypasses the intended hold because `GroupChatWorkspace` checks for a current binding before it checks the retained room's `shippedAdoption`. The descriptor-only binding has no `adoptionOwner`, so `requireRoute()` permits it and `canonicalGroupRequest()` can still dispatch through `host.requestProfile`.

This is directly composable from the new regressions: register the canonical alias first, then hit the newly-fixed source-missing refusal. The current successful-adoption test proves the alias is retired only when `bindAdoptedCanonicalGroup()` is reached; the orphan-refusal test starts with an empty registry, so it does not cover the combined state.

**Required repair:** make the durable adoption checkpoint own/quarantine *all* bindings for that room identity as soon as adoption takes execution ownership, including waiting/prepared/error paths—not only after successful import. Discovery must fail closed/read-only for a retained matching room even when the checkpoint has no route, without reconstructing authority from durable state. Add the combined regression: pre-existing alias → source-missing adoption refusal → old binding no longer current, discovery cannot recreate a writable descriptor alias, opening it cannot emit `groups.send` / member resolution / other mutation; retain a successful-adoption control proving the live owner lease still works.

I re-read the changed production files and the surrounding workspace/request/adoption paths for this head. GitHub still reports this PR open, draft, and non-mergeable. Exact-head hosted queries still return zero workflow runs and no commit statuses, so there is no hosted-green receipt for this object. Current main was observed separately at `130b8f2c5dbca93a81aa396dd2ba44420d78f6f0`; that is not an integration claim for this stacked branch.

apps/desktop/src/plugins/hermes-bots/canonical-group-registry.tsx
```diff
@@ -4,12 +4,158 @@ import { useEffect, useState } from 'react'
...
+
+  if (previous && previous !== routeOwner) {
+    previous.release()
+  }
+
+  const next = { ...$canonicalGroupBindings.get() }
+
+  // Retire any pre-adoption discovery alias as well as the old adopted binding.
+  // Its mounted view retains its old closure and must not become writable again.
+  for (const [alias, existing] of Object.entries(next)) {
```
andrexibiza: **P1 — this alias retirement happens too late for refusal/hold paths.** It only runs after `bindAdoptedCanonicalGroup()` is reached. A pre-existing `canonical:<connection>:<profile>:<roomId>` binding survives when adoption first persists its waiting checkpoint and then the new source-owner validation refuses before import: `persistIssue()` revokes only the retained group key, so this alias stays current. `registerCanonicalGroup()` can also recreate it while the checkpoint is waiting because its retained-adoption lookup requires `adoption.route`. Quarantine every matching binding as soon as `shippedAdoption` takes execution ownership, and regress alias-first → orphan refusal → zero writable canonical mutation.

dokterdok: Fixed in 5211890fb87b626c9ba569910375707530c3a2cc. The checkpoint now quarantines matching bindings before it is published or awaits storage, not only after successful import. Existing hydrated checkpoints invalidate captured descriptor closures as well. Discovery without a resolved route returns the retained read-only room instead of recreating a writable alias; old tabs retain the original history and recovery notice. Known routes retain exact connection/profile scoping, and successful adoption still requires the live owner lease.

The combined alias-first → orphan refusal regression covers three source shapes, rejects Send/member-resolution/removal/disband at the request boundary, prevents alias recreation and verifies read-only rendering. Waiting/prepared/conflict hydration, pending checkpoint persistence, rollback and positive current-owner retry are also covered. Eight causal cases fail on the preceding production; 60 focused tests across eight files pass on this correction, as do renderer TypeScript, scoped ESLint and diff checks.

A fresh task-local authenticated localhost journey using the production SDK and real composed backend passes both the refusal/no-mutation path and the successful lost-ACK/idempotent-retry/history/attachment/new-turn control. This is local composed evidence, not installed/native or public-only acceptance. One independent source-only review found no concrete reachable P1/P2 issue; it did not rerun tests.

The exact PR and remote branch head were read back. Existing runtime-base conflict remains, the PR remains draft, and exact-head hosted queries currently have no workflows/checks/suites/statuses; no hosted-green, merge or deployment claim.

PR #97847, opened 2026-08-29
fix(auth): key global-auth-store and OAuth-heal caches on content digest, not mtime

nftpoetrist:
## Summary

`_load_global_auth_store()` (`hermes_cli/auth.py`) memoises the parsed global-root `auth.json` keyed on `(path, mtime_ns)`. That's the right idiom for a config cache, but this cache guards **credential identity**: an external metadata-preserving copy of the file — a dotfile-sync tool, a backup/restore, anything using `rsync -a` or equivalent — can replace its bytes while restoring the original mtime. A stat-only key never detects that, so the memo keeps serving stale (or entirely different) global credentials for the life of the process, with no TTL to self-heal it.

This is the exact same anti-pattern the Vertex SA credential cache had before `#97701` (merged earlier in this same window) fixed it — same review reasoning, quoted from that fix's own docstring: *"the stat idiom is right for config caches... wrong for a credential cache (guards identity)."* That fix never got applied to this cache.

Reproduced empirically against upstream/main before writing the fix: an atomic-replace with the original `mtime_ns` restored serves stale (and here, entirely different) credential data indefinitely.

Reachability: `_load_global_auth_store()` feeds `read_credential_pool()`/`load_pool()` and xAI OAuth state resolution — normal production paths, not an opt-in/plugin route.

## Fix

Rekey the memo on `(path, sha256-of-content)` instead of `(path, mtime_ns)`, mirroring `agent/vertex_adapter.py`'s `_read_sa_file()` pattern exactly.

## Test plan

- [x] Updated `tests/hermes_cli/test_global_auth_store_memo.py`: the old `test_mtime_change_re_reads_once` only bumped mtime without changing content — under the new key that correctly does NOT force a re-read (nothing changed), so it's rewritten as `test_content_change_re_reads_once`, which changes real content and asserts a re-read.
- [x] Added `test_metadata_preserving_replace_is_not_served_stale` — the direct regression test for this bug: replace the file's content, restore its exact original mtime/atime via `os.utime`, and assert the *new* content is served (not the stale cached one).
- [x] Mutation-verified: stashed the fix, confirmed the new regression test fails against the original mtime-only code — stale `sk-x` served instead of the replaced `sk-DIFFERENT`. Restored the fix; all 4 tests in the file pass.
- [x] Full auth-adjacent suite (`tests/hermes_cli/test_auth*.py`, `test_global_auth_store_memo.py`, `test_xai_oauth_profile_auth.py`, `test_model_flow_pooled_credentials.py`, `test_nous_auth_status_cache.py`, `test_hermes_constants.py` — 239 tests) passes unchanged, 6 pre-existing platform-gated skips.

---

## Update (second commit): same anti-pattern, second cache, same file

While implementing the above, found a **second** cache in this same file with the identical mtime-vs-content-digest anti-pattern: `_heal_forked_single_use_oauth_grants()`'s hot-path short-circuit (`_oauth_heal_clean_marks`, added by `#100929`, merged just before this branch was rebased onto current `upstream/main`). Its fingerprint keyed the "verified clean, skip the locked read-modify-write scan" mark on `(profile auth.json path, mtime_ns, singleton mtime_ns)` — same class of bug, different cache, same file.

A profile's `auth.json` is itself a credential store, so the same external-metadata-preserving-copy scenario applies: a restored/reintroduced OAuth fork would sit unhealed forever once the profile is marked "clean" (no TTL), with `load_pool()` skipping the scan believing nothing changed — silently defeating the entire purpose of this auto-heal mechanism.

**Fix:** rekeyed the fingerprint on sha256-of-content for both files (the profile's `auth.json` and the optional `.anthropic_oauth.json` singleton), mirroring this PR's first fix exactly.

**Tests:** added `test_clean_mark_is_not_fooled_by_a_metadata_preserving_replace` to `tests/agent/test_credential_pool_profile_oauth_fork.py`: heals an existing fork, lets the second call record the clean mark, then replaces the profile's `auth.json` with a freshly forked copy while restoring the exact mtime the mark observed, and asserts the fork is detected and healed rather than served as "clean" forever.

**Mutation-verified**: reverting the fingerprint back to mtime makes the new test fail — `heal_forked_single_use_oauth_grants()` returns `None` (fork left unhealed) instead of a summary dict.

Ran the full `tests/agent/test_credential_pool_profile_oauth_fork.py` suite (16 tests, all pass) plus `test_global_auth_store_memo.py` and `tests/hermes_cli/test_auth*.py` (186 passed, 2 skipped, 1 pre-existing failure — `test_seed_from_singletons_respects_hermes_pkce_suppression` — confirmed independent of this change via `git stash`, fails identically with both commits' changes reverted).

**Competitor check:** searched `gh pr list --state all` for "oauth_heal_clean_marks mtime", "heal_forked_single_use_oauth_grants", "OAuth heal content digest", "forked oauth grant profile cache" — only hit is `#100929` (merged, the PR that introduced this cache in the first place; not a competitor for a fix to its freshness bug). No open or closed PR covers this gap.

### DIFF
--- a/hermes_cli/auth.py
+++ b/hermes_cli/auth.py
@@ -1836,9 +1836,17 @@ _OAUTH_TOKEN_FIELDS = (
 )
 
 _oauth_heal_notices: List[str] = []
-# provider -> (profile auth.json path, auth.json mtime_ns, singleton mtime_ns)
-# of the last store verified fork-free; lets load_pool() skip the locked scan.
-_oauth_heal_clean_marks: Dict[str, Tuple[str, Optional[int], Optional[int]]] = {}
+# provider -> (profile auth.json path, auth.json content digest, singleton
+# content digest) of the last store verified fork-free; lets load_pool() skip
+# the locked scan. Content-digest-keyed, not mtime-keyed: this cache guards
+# whether a credential fork needs healing, and an external metadata-preserving
+# copy (dotfile sync, backup/restore) can replace a forked auth.json's bytes
+# while restoring its original mtime, which a stat-only key would never
+# detect — the mark would keep saying "clean" and the auto-heal (whose whole
+# purpose is fixing exactly this kind of forked grant) would never re-scan
+# (same class of bug as _global_auth_store_cache above and the Vertex SA
+# credential cache — see agent/vertex_adapter.py's _read_sa_file).
+_oauth_heal_clean_marks: Dict[str, Tuple[str, Optional[str], Optional[str]]] = {}
 
 
 def consume_oauth_heal_notices() -> List[str]:
@@ -2000,14 +2008,15 @@ def _heal_forked_single_use_oauth_grants(provider_id: str) -> Optional[Dict[str,
 
     # Hot-path short-circuit: load_pool() runs per model call. Once this
     # profile's store was verified clean for *provider_id*, skip the locked
-    # read-modify-write until the profile's own files change (mtime key).
-    def _stamp(p: Optional[Path]) -> Optional[int]:
+    # read-modify-write until the profile's own files change (content-digest
+    # key — see the type comment above for why not mtime).
+    def _digest(p: Optional[Path]) -> Optional[str]:
         try:
-            return p.stat().st_mtime_ns if p is not None else None
+            return hashlib.sha256(p.read_bytes()).hexdigest() if p is not None else None
         except OSError:
             return None
 
-    fingerprint = (str(profile_path), _stamp(profile_path), _stamp(profile_singleton))
+    fingerprint = (str(profile_path), _digest(profile_path), _digest(profile_singleton))
     if _oauth_heal_clean_marks.get(provider_id) == fingerprint:
         return None
     if fingerprint[1] is None and fingerprint[2] is None:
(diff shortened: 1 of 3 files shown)


Enough1122:

**fix(auth): key the global auth-store memo on content digest, not mtime**

Correct and security-relevant fix. A stat-only key genuinely cannot detect a metadata-preserving replace of `auth.json` (dotfile sync, backup/restore), and serving stale — or entirely different — global credentials indefinitely is a real credential-identity hazard. The digest key closes it, and `test_metadata_preserving_replace_is_not_served_stale` pins the exact scenario (restores both atime and mtime). The parallel to the Vertex SA cache fix is well documented.

Non-blocking observations:

1. **Double read on cache miss.** On a miss the flow reads the file once for the digest, then reads/parses it again for the store. The cost note says digest-read is noise next to the JSON parse, which is fair — but a miss now reads the file twice. If a cheap "read once, hash and parse from the same bytes" refactor is easy, it removes the only inefficiency left.

2. **Full read on every probe.** Every `/model` picker row now does a full `read_bytes()` + SHA-256 of the credentials file before the memo can answer. For a small local auth.json this is fine (and cheaper than the JSON parse, as argued), but worth remembering if the global store ever lands on a slow/remote filesystem — the per-probe cost would stop being negligible. The comment already anticipates this tradeoff; just flagging it explicitly.

PR #97852, opened 2026-08-29
fix(desktop): allow drive_preview on visible session tiles

### LINKED ISSUE #97848: [Bug][Desktop]: In-app browser actions rejected for the visible session
## Bug Description
The Desktop in-app browser can discover and highlight DOM elements, but `drive_preview` actions are rejected with:

`The in-app browser only takes actions in the session the user is looking at.`

The failure occurs even when the browser is displayed in the foreground session. This makes DOM scanning appear to work while typing/clicking fails.

## Steps to Reproduce
1. Open a logged-in web page (observed with ChatGPT) in the Desktop in-app browser.
2. Keep the browser/session visible in the Desktop window.
3. Ask the agent to inspect the page and type into a visible textbox using the in-app browser path.
4. Observe that elements are discovered, but the action is rejected by the session-visibility guard.

## Expected Behavior
Actions should be allowed for the session currently visible to the user, including a focused session tile/runtime, not only the primary active session.

## Actual Behavior
The DOM element inventory is returned, but preview actions can be rejected because the event path uses the generic active-session check rather than the preview-visible session check.

## Environment
- OS: Windows 11
- Surface: Hermes Desktop in-app browser
- Observed page: `https://chatgpt.com/`
- Local repository state at investigation: `c03c72a19`

## Investigation / Proposed Fix
Use one shared visibility predicate for preview open/close/action routing and gateway event handling. A session should count as visible when its ID matches the focused runtime, active session, or a runtime represented by a visible session tile. Preview gateway events should use this preview-specific visibility result instead of the generic `isActiveEvent` flag.

Local validation succeeded with this approach:
- Preview visibility regression tests: passed
- TypeScript/typecheck: passed
- Desktop build: passed
- Real Desktop validation: `drive_preview` discovered the ChatGPT textbox and successfully typed `入力テスト（修正版の内蔵ブラウザ）` without submitting it.

This report is based on one Windows environment; the code path appears general to multi-session/tile state rather than ChatGPT-specific.

astraltrekkin:
## What does this PR do?

`drive_preview` was gated on `isActiveEvent` (routed session === primary `$activeSessionId`). Preview open/close already treat a session as visible when it is the focused runtime, the primary session, or an on-screen session tile. That mismatch rejected click/type in a focused tile even while the in-app browser was in front of the user, while `action="elements"` / read still succeeded.

This change extracts `sessionIsOnScreen` and uses it for preview act as well as open/close, so a visible tile can drive the page and a hidden background session still cannot.

## Related Issue

Fixes #97848

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `apps/desktop/src/lib/preview-visibility.ts` — shared `sessionIsOnScreen` (focused runtime, primary active, or tile `runtimeId`)
- `apps/desktop/src/app/session/hooks/use-preview-routing.ts` — preview open/close use the shared helper
- `apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/desktop-bridge.ts` — `preview.act.request` uses `sessionIsOnScreen` instead of `isActiveEvent`
- Tests: `preview-visibility.test.ts`, `desktop-bridge.test.ts`

## How to Test

1. Open Hermes Desktop with two sessions on screen (primary chat plus a session tile), and open a page in the in-app browser from the tile session.
2. From the tile session, ask the agent to list elements and type into a visible field via `drive_preview`.
3. Confirm the action runs instead of `The in-app browser only takes actions in the session the user is looking at.`
4. From a session that is not on screen, confirm the same action is still rejected.
5. From `apps/desktop`: `npx vitest run src/lib/preview-visibility.test.ts src/app/session/hooks/use-message-stream/gateway-event/desktop-bridge.test.ts`

### Code

- [ ] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [x] I've tested on my platform: macOS 26.5

## Screenshots / Logs

```

```

### DIFF
--- a/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/desktop-bridge.test.ts
+++ b/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/desktop-bridge.test.ts
@@ -64,7 +64,12 @@ describe('handleDesktopBridgeEvent preview.act visibility', () => {
     handleDesktopBridgeEvent(actCtx('runtime-tile', false))
     await vi.waitFor(() => expect(request).toHaveBeenCalled())
 
-    const [, params] = request.mock.calls[0] as [string, { text: string }]
-    expect(params.text).not.toContain('only takes actions in the session the user is looking at')
+    expect(request).toHaveBeenCalledWith(
+      'preview.act.respond',
+      expect.objectContaining({
+        request_id: 'req-97848',
+        text: expect.not.stringContaining('only takes actions in the session the user is looking at')
+      })
+    )
   })
 })
--- a/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/desktop-bridge.test.ts
+++ b/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/desktop-bridge.test.ts
@@ -1,0 +1,70 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+vi.mock('@/app/chat/right-rail/preview-act', () => ({
+  actOnActivePreview: vi.fn(async () => ({ success: true, typed: 'hello' }))
+}))
+
+import { $gateway } from '@/store/gateway'
+import { $activeSessionId } from '@/store/session'
+import { $sessionTiles, clearAllSessionStates } from '@/store/session-states'
+
+import { handleDesktopBridgeEvent } from './desktop-bridge'
+import type { GatewayEventContext } from './types'
+
+function actCtx(sessionId: string, isActiveEvent: boolean): GatewayEventContext {
+  return {
+    deps: {} as GatewayEventContext['deps'],
+    event: { session_id: sessionId, type: 'preview.act.request' },
+    explicitSid: sessionId,
+    fromActiveSource: () => true,
+    isActiveEvent,
+    occurredAt: Date.now() / 1000,
+    payload: { action: 'type', ref: 'inp-prompt', request_id: 'req-97848', text: 'hello' },
+    scheduleConfigRefresh: () => undefined,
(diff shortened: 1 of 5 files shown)


Enough1122:

**fix(desktop): allow drive_preview on visible session tiles**

Clean fix. Extracting `sessionIsOnScreen` into a shared `lib/preview-visibility.ts` and using it for the `preview.act.request` gate (alongside preview open/close) removes the primary-only `isActiveEvent` restriction while keeping hidden background sessions blocked. Good unit coverage in `preview-visibility.test.ts` and `desktop-bridge.test.ts`, including the "visible tile that is not primary" case.

Non-blocking observations:

1. **Redundant guard.** In `desktop-bridge.ts`, `const previewVisible = !!sessionId && sessionIsOnScreen(sessionId)` — `sessionIsOnScreen` already returns `false` for an empty/falsy `sessionId` (see `if (!sessionId) return false`). The `!!sessionId &&` is redundant; harmless, but could be dropped for clarity.

2. **Gate-only test coverage.** The `desktop-bridge.test.ts` mocks `@/app/chat/right-rail/preview-act` (and the real path goes through the dynamically imported `loadPreviewEngine`), so the tests assert only the visibility gate — not that the preview action actually executes against the visible tile. That's the right scope for this fix (the gate is the change), just noting execution itself isn't exercised.

3. **Behavioral breadth.** The gate moved from "primary active event only" to "any on-screen session (focused runtime, primary, or tile)". If a future layout allows two tiles visible simultaneously, both could drive the shared in-app preview page. Currently tiles are effectively single, so low risk — flagging for awareness.

PR #47218, opened 2026-06-16
feat(whatsapp): wire channel prompts into the WhatsApp adapter

JiraiyaETH:
Upstream supports per-channel prompts (`resolve_channel_prompt` in `gateway/platforms/base.py`) and bridges `channel_prompts` config for all platforms — but the WhatsApp adapter never calls `resolve_channel_prompt` or passes `channel_prompt` to `MessageEvent`, unlike Slack/Telegram/Discord/Mattermost. This wires it up, closing the gap.

PR #29883, opened 2026-05-21
feat(whatsapp): support channel_prompts for per-user/per-group prompts

DRStranglove:
## Summary

WhatsApp adapter was the only gateway platform not consuming the shared `resolve_channel_prompt()` helper. Slack, Mattermost, and Telegram all inject per-channel ephemeral prompts from `channel_prompts` in their platform config; WhatsApp ignored them.

This wires WhatsApp into the same pattern.

## Motivation

Family/multi-user WhatsApp bots: each contact (or group) should be able to get a distinct persona — different language, tone, allowed capability set — without running a separate gateway process and a separate phone number per persona.

Without this, every WhatsApp DM/group sees the same global system prompt regardless of who's writing.

## Behavior

| Scenario | Lookup key | Fallback |
|---|---|---|
| Direct message | `senderId` | `chatId` |
| Group message | `chatId` only | — (sender's personal prompt does NOT leak into the group) |

Matches the existing Slack/Telegram/Discord semantics. Blank/whitespace prompts are treated as absent (handled by `resolve_channel_prompt` already).

## Config example

```yaml
whatsapp:
  channel_prompts:
    '436XXXXXXXXX@s.whatsapp.net': 'You are X. Reply in Spanish.'
    '436YYYYYYYYY@s.whatsapp.net': 'You are kid-safe assistant. Reply in ES or DE only.'
    '120XXXXXXXXX@g.us':           'Family group assistant.'
```

## Changes

- `gateway/platforms/whatsapp.py` — +11 lines: import the helper, derive `_primary_key` from `isGroup`, pass `channel_prompt` to `MessageEvent`.
- `tests/gateway/test_whatsapp_channel_prompts.py` — new, 7 tests covering DM, group, fallback, no-match, whitespace, and the group-doesn't-match-sender invariant.

## Test results

```
$ python -m pytest tests/gateway/test_whatsapp_*.py -q
94 passed, 1 warning in 4.63s
```

Existing channel-prompt tests for other platforms still pass:

```
$ python -m pytest tests/gateway/test_discord_channel_prompts.py \
                   tests/gateway/test_telegram_group_gating.py \
                   tests/gateway/test_config.py \
                   tests/hermes_cli/test_config.py \
                   tests/gateway/test_pending_event_none.py -q
154 passed in 3.91s
```

## Backward compatibility

[...]

PR #22209, opened 2026-05-09
feat(gateway): add channel_prompt and auto_skill resolution for WhatsApp adapter

hermes-matekno:
## What does this PR do?

Adds `channel_prompt` and `auto_skill` (channel_skill_bindings) resolution to the WhatsApp platform adapter, matching the existing behavior in Discord, Telegram, and Slack adapters.

Currently, when a message arrives from a WhatsApp group, the `_build_message_event()` method in `whatsapp.py` creates a `MessageEvent` without populating `channel_prompt` or `auto_skill`. This means per-channel context injection and skill auto-loading — features already working for Discord and Telegram — are completely ignored for WhatsApp.

### `gateway/platforms/whatsapp.py`
- In `_build_message_event()`: resolve `chatId` from message data, call `resolve_channel_prompt()` and `resolve_channel_skills()` from `gateway.platforms.base`, and pass the results as `channel_prompt=` and `auto_skill=` to `MessageEvent`

### `gateway/config.py`
- Extended the `channel_skill_bindings` config bridge from `(Platform.DISCORD, Platform.SLACK)` to `(Platform.DISCORD, Platform.SLACK, Platform.WHATSAPP)` so WhatsApp platform configs can specify skill bindings per channel

## How to test

1. Add a WhatsApp group to `config.yaml`:
```yaml
whatsapp:
  channel_prompts:
    "120363427088953130@g.us": |
      This is the birthday reminders group. Always use the birthday-tracker skill.
  channel_skill_bindings:
    - id: "120363427088953130@g.us"
      skills: ["birthday-tracker"]
```

2. Restart the gateway (`hermes gateway restart`)
3. Send a message from that WhatsApp group
4. Verify the agent loads the specified skill and follows the channel prompt context

## Type of change

- [x] New feature (non-breaking change adding functionality)

## Checklist

- [x] I have read the CONTRIBUTING.md document
- [x] My commit messages follow Conventional Commits
- [x] I have searched for duplicate PRs
- [x] Only related changes are included
- [ ] Tests added for this change
- [x] Cross-platform compatibility considered (WhatsApp only, reuses existing cross-platform infrastructure)
- [x] Config example updated (channel_skill_bindings now accepted for WhatsApp)

PR #97853, opened 2026-08-29
feat(whatsapp): per-channel system prompt (channel_prompt) for WhatsApp adapters

kk17:
Add per-channel ephemeral system prompts (channel_prompt) to the WhatsApp adapters, mirroring the pattern already implemented for Discord/Feishu/Telegram/Slack.

## What
- `gateway/platforms/whatsapp_cloud.py`: resolve `channel_prompt` via `resolve_channel_prompt(config.extra, chat_id)` and pass to `MessageEvent`.
- `plugins/platforms/whatsapp/adapter.py` (Baileys): same, keyed off `chatId`.

## Why
WhatApp was the only major platform adapter without channel-prompt support. The legacy `channel_prompts` config path is explicitly supported by upstream (`gateway/run.py` applies `event.channel_prompt`).

## Related
- #47218, #22209, #29883 (earlier similar efforts — this one rebased onto current main)
- `resolve_channel_prompt` exists in `gateway/platforms/base.py`

### DIFF
--- a/gateway/platforms/whatsapp_cloud.py
+++ b/gateway/platforms/whatsapp_cloud.py
@@ -2099,6 +2099,11 @@ class WhatsAppCloudAdapter(WhatsAppBehaviorMixin, BasePlatformAdapter):
             if body:
                 rich_sent_store.record(chat_id, wamid, body)
 
+        # Per-channel ephemeral prompt (mirrors Telegram/Slack adapters).
+        # See https://github.com/NousResearch/hermes-agent/pull/47218
+        from gateway.platforms.base import resolve_channel_prompt
+        _channel_prompt = resolve_channel_prompt(self.config.extra, chat_id)
+
         return MessageEvent(
             text=body,
             message_type=message_type,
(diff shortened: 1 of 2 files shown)


Enough1122:

**feat(whatsapp): per-channel system prompt (channel_prompt) for WhatsApp adapters**

Low-risk feature that mirrors the existing Telegram/Slack `resolve_channel_prompt` pattern in both WhatsApp adapters (`gateway/platforms/whatsapp_cloud.py` and `plugins/platforms/whatsapp/adapter.py`). The lazy import inside the function body is a reasonable circular-import guard, and attaching the resolved prompt to `MessageEvent` is consistent with how the other adapters wire it.

Non-blocking observations:

1. **No test coverage.** Neither adapter change has a test asserting that `channel_prompt` is resolved from `config.extra` and attached to the `MessageEvent`. Given the fix is a wiring change (easy to regress), consider a small test mirroring whatever coverage the Telegram/Slack adapters have for this path.

2. **Per-message resolution cost.** `resolve_channel_prompt(self.config.extra, chat_id)` runs on every inbound message build. If it's a cheap config read, fine — but if it does any file I/O or parsing per message, it's a hot-path cost worth checking once.

3. **Trivial doc nit.** In `plugins/platforms/whatsapp/adapter.py`, the removed line "metadata + text tagging are unconditional when the flag is present" leaves the comment reading "...at the bridge layer; so a future producer can set it without adapter changes." — the dangling semicolon reads awkwardly. Cosmetic only.

PR #63289, opened 2026-07-12
fix(whatsapp): degrade flapping bridge health

LeonSGP43:
## What does this PR do?

This makes WhatsApp bridge flapping observable and recoverable. The bridge now reports sustained reconnect loops as `status: "degraded"` in `/health`, and the Python WhatsApp adapter treats that degraded health as a retryable fatal error so Hermes can restart or reconnect the platform instead of silently staying "connected" while inbound messages are dropped.

## Related Issue

Fixes #63277

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Added rolling disconnect tracking in [`scripts/whatsapp-bridge/bridge.js`](scripts/whatsapp-bridge/bridge.js) so `/health` returns `degraded` plus reconnect counters during sustained Baileys flap loops.
- Added adapter-side degraded-health handling in [`plugins/platforms/whatsapp/adapter.py`](plugins/platforms/whatsapp/adapter.py) so `_poll_messages()` escalates flapping into the existing retryable fatal/reconnect path.
- Added a regression test in [`tests/gateway/test_whatsapp_connect.py`](tests/gateway/test_whatsapp_connect.py) covering degraded bridge health.

## How to Test

1. Run `node --check scripts/whatsapp-bridge/bridge.js`.
2. Run `uv run pytest tests/gateway/test_whatsapp_connect.py -q`.
3. Reproduce a WhatsApp reconnect loop and confirm `/health` reports `status: "degraded"`, then verify Hermes queues WhatsApp for reconnect instead of leaving it silently connected.

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
- [x] My PR contains **only** changes related to this fix/feature (no unrelated commits)
- [ ] I've run `pytest tests/ -q` and all tests pass
[...]

PR #97854, opened 2026-08-29
fix(whatsapp): detect flapping bridge health and surface as retryable fatal error

kk17:
Detect sustained WhatsApp bridge connection flapping (process alive but repeatedly disconnecting/reconnecting) and surface it as a retryable fatal error, instead of waiting for the bridge process to fully exit.

## What
- `plugins/platforms/whatsapp/adapter.py`: add `_check_bridge_health_degraded()` polling bridge `/health`; when `status == degraded`, set retryable fatal error.
- `scripts/whatsapp-bridge/bridge.js`: track `disconnectTimestamps[]` (rolling 60s window, >5 disconnects = degraded), expose via `buildHealthSnapshot()` in `/health`.
- Tests: adapter-side degraded-payload test.

## Why
Current upstream only detects a *terminated* bridge process. A bridge that stays alive but flaps silently wedges without any alert — this was the root cause of a real multi-day silent WhatsApp outage.

## Notes
Based on earlier work in #63289 (by LeonSGP43); this version is rebased onto current main and includes the adapter test.

### DIFF
--- a/plugins/platforms/whatsapp/adapter.py
+++ b/plugins/platforms/whatsapp/adapter.py
@@ -876,6 +876,52 @@ class WhatsAppAdapter(WhatsAppBehaviorMixin, BasePlatformAdapter):
             await self._notify_fatal_error()
         return self.fatal_error_message or message
 
+    async def _check_bridge_health_degraded(self) -> Optional[str]:
+        """Surface sustained bridge flapping as a retryable fatal error."""
+        if not self._http_session:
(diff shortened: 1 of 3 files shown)


alt-glitch:

Duplicate of #63289. Both PRs implement the same bridge degraded-health signal, adapter retryable-fatal handling, and regression coverage; #63289 is the earlier open canonical PR.

PR #97855, opened 2026-08-29
fix(peer): keep remote DM turns alive over SSE

### LINKED ISSUE #97769: Bug: `hermes peer dm` times out against Windows peer gateway — http.client vs aiohttp /chat incompatibility
### Environment (anonymized)
- **Controller (host-a):** macOS 15, Hermes Agent v0.20.6 (upstream `c03c72a1`), Python 3.11.15 (venv)
- **Peer (host-b):** Windows 11, Hermes Agent v0.20.6 (fresh install), running `hermes gateway run` on port 8642; gateway is aiohttp/3.14.3
- Both hosts on the same LAN subnet; TCP connectivity between them verified.
- Network is plain HTTP (no TLS) between the two hosts.

### Summary
`hermes peer dm <peer> "<message>"` from host-a to host-b's gateway always fails with:
```
Could not reach peer '<peer>': timed out
```
Layer-by-layer investigation on host-a shows the failure is **not** network, auth, model config, or routing — it is a client/server transport incompatibility between Python's stdlib `http.client` (used by `peer dm`) and the peer gateway's aiohttp `/chat` endpoint.

### What works (proves the gateway, network, key, and model are all fine)
On host-a:
- `curl -X POST http://<peer-ip>:8642/api/sessions/<id>/chat -H "Authorization: Bearer <key>" -d '{"message":"hi"}'` → **HTTP 200, full JSON reply in ~4s**.
- `curl http://<peer-ip>:8642/api/sessions` (GET/POST) → 200, fast.
- `curl http://<peer-ip>:8642/v1/chat/completions` → 200, model returns a reply.
- Raw `socket.create_connection(("<peer-ip>",8642), timeout=10)` from the **same venv python** → succeeds.
- On the peer itself (host-b, PowerShell `Invoke-RestMethod` to `127.0.0.1:8642/api/sessions/<id>/chat`) → **200, reply in ~3.7s**.

### What fails
Any Python client built on `http.client` on host-a, against the **same** `POST /api/sessions/<id>/chat` endpoint:
- `hermes peer dm` (stdlib `urllib.request.urlopen`)
- a minimal `urllib.request` reproduction of `peer.py::_request`
- `requests`/`urllib3`

All fail identically at `http.client` `response.begin()` → `_read_status()` → `self.fp.readline(...)` with `TimeoutError: timed out`. The TCP connection is established (verified via `netstat` — state `ESTABLISHED`), but the client never receives the HTTP status line. `Connection: close` header does not help. Forcing IPv4 (`socket.AF_INET`) does not help.

### Root-cause hypothesis
The peer gateway (aiohttp) serves `POST /api/sessions/<id>/chat` as a synchronous agent turn. While libcurl (curl) reads the response correctly, Python's `http.client` blocks forever waiting for the status line. This strongly suggests a server-side flush/backpressure/keep-alive interaction in aiohttp that the stdlib client does not tolerate — i.e. a transport-layer mismatch specific to this endpoint (other endpoints on the same gateway respond fine to stdlib clients).

### Workaround that succeeds
Wrapping the exact `peer dm` flow (find/create `BOT_MODE_CHAT` session + `POST /chat` + print reply) in a `curl`-based script works reliably and returns the peer's reply. So the peer protocol itself is fine; only the stdlib HTTP client in `peer.py` is affected.

### Minimal reproduction (host-a, anonymized)
```python
import os, json, urllib.request
KEY = os.environ["PEER_KEY"]                       # the peer's API_SERVER_KEY
BASE = "http://192.0.2.10:8642"                    # host-b gateway (placeholder IP)
# create a bot chat session
h = urllib.request.Request(f"{BASE}/api/sessions",
    data=json.dumps({"title":"BOT_MODE_CHAT","source":"bot_peer_dm"}).encode(),
    method="POST", headers={"Authorization":f"Bearer {KEY}","Content-Type":"application/json"})
sid = json.loads(urllib.request.urlopen(h, timeout=20).read())["session"]["id"]
# this call TIMES OUT even though curl to the identical URL returns in ~4s:
c = urllib.request.Request(f"{BASE}/api/sessions/{sid}/chat",
    data=json.dumps({"message":"hi"}).encode(), method="POST",
    headers={"Authorization":f"Bearer {KEY}","Content-Type":"application/json"})
urllib.request.urlopen(c, timeout=120).read()      # -> TimeoutError at _read_status
```

### Expected
`hermes peer dm` should return the peer's reply (as it does via curl / via the peer's own local client).

### Actual
`Could not reach peer '<peer>': timed out` (urllib `TimeoutError` at `http.client` `_read_status`).

### Suggested fix direction
In `hermes_cli/subcommands/peer.py::_request`, the stdlib `urllib.request` client appears to hang on the `/chat` endpoint against an aiohttp gateway. Consider using a robust HTTP client (e.g. the same aiohttp/httpx client the agent core uses elsewhere) for peer dispatch, or ensure the gateway flushes/completes the `/chat` response in a way stdlib `http.client` can consume. Happy to test a patch on the macOS controller + Windows peer setup.

rakeshr:
## Summary

- Route `hermes peer dm` agent turns through the existing `/api/sessions/{id}/chat/stream` endpoint.
- Parse the terminal `assistant.completed` SSE event into the existing CLI result shape.
- Preserve the existing peer authentication, Bot Chat lookup/creation, JSON output, and error handling.
- Add a loopback regression test that emits an SSE keepalive and completion event and rejects the buffered `/chat` route.

Fixes #97769

## Why

The buffered `/chat` endpoint does not send its HTTP response until the full agent turn finishes. The SSE endpoint prepares headers before the turn and emits keepalives, avoiding idle/status-line timeouts observed when `hermes peer dm` talks to a Windows gateway.

## Tests

- `PYTHONPATH=. uv run --no-project --with pytest pytest --confcutdir=tests/hermes_cli tests/hermes_cli/test_peer_cmd.py -q`
  - 14 passed
- `uv run --no-project --with ruff ruff check hermes_cli/subcommands/peer.py tests/hermes_cli/test_peer_cmd.py`
  - All checks passed
- `uv run --no-project python -m py_compile hermes_cli/subcommands/peer.py tests/hermes_cli/test_peer_cmd.py`
  - Passed
- `git diff --check`
  - Passed

The focused regression test was also run against the old buffered route and failed 4 tests, confirming the test detects the regression.

### DIFF
--- a/hermes_cli/subcommands/peer.py
+++ b/hermes_cli/subcommands/peer.py
@@ -11,8 +11,8 @@ bot on THIS machine a transport to message bots on THAT machine:
     hermes peer status spark run_abc123
 
 ``dm`` resolves the remote agent's canonical "Bot Chat" session (by title,
-creating it when missing), runs ONE synchronous agent turn over the peer's
-existing ``POST /api/sessions/{id}/chat`` endpoint, and prints the reply on
+creating it when missing), runs ONE agent turn over the peer's existing
+``POST /api/sessions/{id}/chat/stream`` endpoint, and prints the reply on
 stdout — the exact cross-machine twin of the local
 ``hermes -p <bot> chat --in ~ -c "Bot Chat" ...`` bot-messaging command, so
 the Bot Mode protocol composes over it unchanged.
@@ -121,6 +121,52 @@ def _request(
         raise RuntimeError("Peer returned a non-object JSON response")
     return parsed
 
+
+def _request_chat_stream(url: str, key: str, message: str) -> dict:
+    """Run one peer chat turn and project its terminal SSE event."""
+    req = urllib.request.Request(
+        url,
+        data=json.dumps({"message": message}).encode("utf-8"),
+        method="POST",
+        headers={
+            "Authorization": f"Bearer {key}",
+            "Content-Type": "application/json",
+            "Accept": "text/event-stream",
+            "User-Agent": "hermes-peer-dm",
+        },
+    )
+    completed = None
+    event = ""
+    data_lines: list[str] = []
+
+    with urllib.request.urlopen(req, timeout=DM_TIMEOUT_S) as resp:  # noqa: S310 — user-registered peer URL
+        for raw_line in resp:
+            line = raw_line.decode("utf-8", "replace").rstrip("\r\n")
+            if line.startswith("event:"):
+                event = line[len("event:") :].strip()
+            elif line.startswith("data:"):
+                data_lines.append(line[len("data:") :].lstrip())
+            elif not line:
+                if event in {"assistant.completed", "error"}:
+                    payload_text = "\n".join(data_lines)
+                    try:
+                        payload = json.loads(payload_text)
+                    except ValueError as exc:
+                        raise RuntimeError(f"Peer returned invalid {event} event: {payload_text[:200]}") from exc
+                    if not isinstance(payload, dict):
+                        raise RuntimeError(f"Peer returned a non-object {event} event")
+                    if event == "error":
+                        raise RuntimeError(str(payload.get("message") or "Peer chat failed"))
+                    completed = payload
+                event = ""
+                data_lines = []
+
+    if completed is None:
+        raise RuntimeError("Peer chat stream ended without a completed assistant response")
+    return {
+        "session_id": completed.get("session_id"),
+        "message": {"role": "assistant", "content": completed.get("content") or ""},
+    }
 
 def _base_url(peer: dict, profile: str | None) -> str:
     url = str(peer.get("url") or "").rstrip("/")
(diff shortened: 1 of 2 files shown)


Enough1122:

**fix(peer): keep remote DM turns alive over SSE**

Sound switch from the buffered `/chat` route to `/chat/stream` (SSE), which is the right fix for idle-timeout drops on long peer turns: the streaming twin prepares headers immediately and emits keepalives. The hand-rolled SSE parser is careful — it accumulates `data:` lines per event, joins with `\n`, dispatches on `assistant.completed`/`error`, and resets state on blank lines. The projection back into the old `_request` result shape (`session_id` + `message.role/content`) keeps callers unchanged, and the test correctly forces the buffered route to fail-closed (500) so a regression can't silently switch back.

Non-blocking observations:

1. **The test doesn't exercise real streaming.** The fake server writes the entire SSE payload (keepalive + `assistant.completed` + `done`) in a single write with `Content-Length`, so the client likely receives it in one read. The parser is validated, but the actual property this PR targets — events arriving incrementally over time with the socket otherwise idle — is not. Consider a test that writes the keepalive, then the completed event in a second write after a small delay, to emulate incremental delivery (and confirm the socket timeout doesn't fire between frames).

2. **`DM_TIMEOUT_S` is now a per-read socket timeout, not a total deadline.** With keepalives arriving inside the window, a single peer turn can run well past `DM_TIMEOUT_S` wall-clock. That's the intended idle-timeout survival, but worth confirming the CLI UX is acceptable for a turn that can now take arbitrarily long.

3. **Empty-data edge.** If the stream ends on a `done` event without ever dispatching `assistant.completed`, `completed` stays `None` and the function raises "ended without a completed assistant response" — a clear, correct failure mode. No issue, just noting the contract.

PR #72673, opened 2026-07-27
fix(qqbot): notify gateway watcher on reconnect exhaustion instead of silent deathFix/qqbot notify gateway

cadezhou:
The QQ (`qqbot`) adapter died silently once its WebSocket reconnect attempts were exhausted (it just called `_mark_disconnected()`), leaving the listener task dead with nothing watching it — the platform never recovered.

This PR makes all three exhaustion paths in `_listen_loop` (rate-limited 4008, `QQCloseError` backoff, generic `Exception`):
- set a `qq_reconnect_exhausted` fatal error (`retryable=True`) so the gateway reconnect watcher takes over;
- schedule the notify as a detached task via `_schedule_fatal_notify()` instead of awaiting inline — the notify chain calls `disconnect()`, which cancels the running `_listen_task`, so an inline await would abort the gateway handler before it re-enqueues the adapter.

Tests
-----
* Added `TestReconnectExhaustionHandoff` covering all three paths, asserting a retryable `qq_reconnect_exhausted` is set, `_mark_disconnected()` is no longer called, and the notify is scheduled as a detached task.
* `tests/gateway/test_qqbot.py` passes 169/169.

Closes #29005

PR #19414, opened 2026-05-03
fix(qqbot): fix 5 reconnect bugs — zombie state, lost close codes, no cooldown, heartbeat timing, missing heartbeat task

Allonz:
## Problem

QQBot adapter enters a **zombie state** after repeated WebSocket disconnections — the adapter process remains alive but stops receiving messages. This was observed in production logs where the connection dropped (code 4009: Session timed out) and then went silent for 93+ minutes.

Root cause: when `_listen_loop` exhausts its reconnect attempts, it simply `return`s without notifying the Gateway's `_platform_reconnect_watcher`. The watcher never knows the adapter has given up, so no higher-level reconnection is attempted.

---

### P0 — Zombie state: `_listen_loop` exit silently dies without notifying Gateway

**Symptom:** After `MAX_RECONNECT_ATTEMPTS` reconnect failures, `_listen_loop` returns but the adapter process stays alive. No messages are received. The `_platform_reconnect_watcher` in `gateway/run.py` is unaware the adapter has given up.

**Fix:** All three exit paths in `_listen_loop` now call `_set_fatal_error("qq_reconnect_exhausted", ..., retryable=True)` before returning. This signals the Gateway's reconnect watcher to take over reconnection at the platform management level.

**Affected paths:**
- Rate limit 4008 backoff exhaustion
- `QQCloseError` (CLOSE/CLOSED events) exhaustion
- Generic `Exception` exhaustion

### P1 — CLOSED/ERROR WebSocket events lose close code and reason

**Symptom:** `WSMsgType.CLOSED` and `WSMsgType.ERROR` events raised a plain `RuntimeError("WebSocket closed")` with no close code or reason. This prevented proper error classification (e.g., distinguishing a server-side 4009 Session Timeout from a network error).

**Fix:** Changed to raise `QQCloseError(msg.data, msg.extra)` instead, preserving the close code and reason for downstream error classification logic.

### P1 — No cooldown period between reconnect failures

**Symptom:** After a failed reconnect attempt, the adapter immediately tries again. The QQ server may not have finished cleaning up the old session, causing the new attempt to fail with a session conflict.

**Fix:** Added a 15-second `await asyncio.sleep(15)` in `_reconnect()` after a failed connection attempt, giving the server time to clean up the old session before the next retry.

### P2 — Heartbeat interval reset on reconnect failure

[...]

PR #17814, opened 2026-04-30
fix(qqbot): trigger fatal error + reconnect when max reconnect attempts reached

fengtianyu88:
## Summary

When the QQ adapters _listen_loop() exhausts all reconnect attempts (MAX_RECONNECT_ATTEMPTS=100), it previously returned silently without notifying the gateway. This left the bot in a zombie state: the gateway believed the platform was connected (is_connected returned True) while the adapter was dead.

## Fix

Use the same _set_fatal_error() + _notify_fatal_error() pattern that telegram.py uses. This triggers:

1. _handle_adapter_fatal_error() in the gateway run loop
2. Adapter removed from self.adapters
3. Platform added to _failed_platforms for background reconnection
4. _platform_reconnect_watcher() picks it up and retries with exponential backoff

Three paths are covered:
- Rate limit (code 4008) exhausts retries
- QQCloseError reconnect failures exceed limit
- General WebSocket exception reconnect failures exceed limit

## Root Cause

Analyzed 11 Reconnect failed events in errors.log:
- All 11 occurred after hermes-gateway restarts
- Pattern: gateway crashes -> adapter is stopped -> REST API calls fail -> gateway restarted -> adapter reconnects
- Fix ensures adapter proactively signals fatal error so gateway reconnects automatically without manual restart

PR #97857, opened 2026-08-29
fix(qqbot): notify gateway when the listen loop gives up instead of going zombie

ttxs69:
## Problem

After a network outage longer than the reconnect budget (backoff 2s→5s→10s→30s→60s × 100 attempts, roughly 1–2h of continuous failure), the QQBot platform adapter becomes a permanent zombie:

- `_listen_loop` exhausts `MAX_RECONNECT_ATTEMPTS` and exits via `_mark_disconnected(); return` — **without setting a fatal error and without notifying the gateway**
- The `_failed_platforms` reconnect watcher only covers platforms that failed at gateway **startup**; nothing polls installed adapters post-startup, and the zombie stays installed in `self.adapters` with `is_connected=False` forever
- Inbound events are gone (no WebSocket); every send waits 15s in `_wait_for_reconnection()` then fails with `Not connected`
- Only a gateway restart recovers the platform

Real-world instance: a DNS outage on 2026-08-19 (~2h, `[Errno 8]`) left a production gateway's qqbot dead for 10 days while Telegram self-healed via the watcher. Gateway logs show the exact sequence: `Reconnect failed: Failed to get QQ Bot access token` ×100 → `Max reconnect attempts reached` → silence.

The same silent-death pattern existed in three more exits:

- **quick-disconnect** (×3 rapid disconnects) and **fatal close codes** (banned/offline/etc.) did call `_set_fatal_error()` but never notified — nothing reads the flag of an installed adapter, so it was equally zombie
- **4008 rate-limit exhaustion** also exited via bare `_mark_disconnected(); return`

## Fix

Every terminal `_listen_loop` exit now goes through `_give_up_and_notify(code, message, retryable=)`:

1. `_set_fatal_error(...)` — error code + retryable flag, `_running=False`
2. `_mark_disconnected()` — `is_connected=False` (ordering matters: it skips its redundant status write once `has_fatal_error` is set)
3. `_notify_fatal_error_detached()` — schedules the gateway's `_handle_adapter_fatal_error` on its own task

The gateway handler then disconnects the zombie and requeues the platform for background reconnection (retryable: max-reconnect ×3, quick-disconnect) or marks it fatal (non-retryable: banned/offline/…) — the exact contract every other adapter already uses; the qqbot listen loop was the one place that never called it.

**Why detached?** The gateway handler calls `adapter.disconnect()`, which awaits `self._listen_task`. Awaiting that from inside the listen loop itself would deadlock (task awaiting itself). The detached task runs the notification after the listen loop has already returned, so the cancel/await in `disconnect()` completes instantly. The wrapper is exception-guarded (best-effort notify) so a gateway-side bug cannot leave an unawaited-task error, and a strong reference on the adapter prevents mid-flight GC.

Two secondary fixes folded in:

- The quick-disconnect exit returns before the shared `_fail_pending("Connection closed")` — it now fails pending response futures itself (defensive parity with sibling adapters)
- Fatal close-code error codes are slugified (`qq_invalid_opcode`, `qq_offline_sandbox_only`) since they now persist in runtime-status `error_code` fields

## Behavior matrix

| Exit | Error code | Retryable | Gateway action |
|---|---|---|---|
| Max reconnect ×3 (generic / QQCloseError / rate-limit) | `qq_max_reconnect` | ✅ | disconnect + requeue, auto-recover when network returns |
| Quick-disconnect ×3 | `qq_quick_disconnect` | ✅ | requeue; recovers once credentials/permissions are fixed |
| Fatal close codes (4001/4002/4010–4014/4914/4915) | `qq_<desc>` (slugified) | ❌ | mark fatal — retrying a banned bot is pointless |

## Tests

4 regression tests in `TestListenLoopFatalNotify` (full suite: 67 passed):

- max-reconnect exhaustion → retryable fatal + handler notified, with the handler calling `adapter.disconnect()` — a regression to inline notification would deadlock/timeout here
- fatal close code (4915) → non-retryable + notified
- quick-disconnect → pending futures settled (future injected late + message asserted, so the assertion discriminates the branch-local `_fail_pending` from the shared one; verified by mutation test)
- raising gateway handler → exception swallowed, no unawaited task error

## Notes

- No gateway-side changes needed — the reconnect watcher, stale-instance guard, and shutdown-race handling (`shutdown_event` check) already exist upstream; this PR just makes qqbot use them
- Reconnect parameters unchanged — the budget is reasonable; only the death behavior after exhausting it was wrong
- No linked issue: searched open PRs/issues for qqbot reconnect/zombie/not-connected, nothing matches

### DIFF
--- a/gateway/platforms/qqbot/adapter.py
+++ b/gateway/platforms/qqbot/adapter.py
@@ -251,6 +251,7 @@ class QQAdapter(BasePlatformAdapter):
         self._http_client: Optional[httpx.AsyncClient] = None
         self._listen_task: Optional[asyncio.Task] = None
         self._heartbeat_task: Optional[asyncio.Task] = None
+        self._fatal_notify_task: Optional[asyncio.Task] = None
         self._heartbeat_interval: float = 30.0  # seconds, updated by Hello
         self._session_id: Optional[str] = None
         self._last_seq: Optional[int] = None
@@ -515,6 +516,49 @@ class QQAdapter(BasePlatformAdapter):
         )
         logger.info("[%s] WebSocket connected to %s", self._log_tag, gateway_url)
 
+    async def _notify_fatal_error_guarded(self) -> None:
+        """Best-effort fatal-error notification — never raises.
+
+        Runs on its own task; an escaping exception would surface only as
+        "Task exception was never retrieved" at GC time and the gateway
+        would silently never learn about the dead adapter.
+        """
+        try:
+            await self._notify_fatal_error()
+        except Exception:  # noqa: BLE001 - notification is best-effort
+            logger.debug("[%s] fatal-error notify failed", self._log_tag, exc_info=True)
+
+    def _notify_fatal_error_detached(self) -> None:
+        """Schedule gateway notification of a fatal error without awaiting it.
+
+        The gateway's fatal-error handler calls ``adapter.disconnect()``,
+        which awaits ``self._listen_task`` — awaiting this very task from
+        inside the listen loop itself would deadlock, so the notification
+        runs on its own task. A strong reference is kept so the task is not
+        garbage-collected mid-notification.
+        """
+        try:
+            self._fatal_notify_task = asyncio.create_task(
+                self._notify_fatal_error_guarded()
+            )
+        except RuntimeError:
+            # No running event loop (e.g. interpreter shutdown) — nothing to notify.
+            pass
+
+    def _give_up_and_notify(self, code: str, message: str, *, retryable: bool) -> None:
+        """Terminal listen-loop exit: set a fatal error and notify the gateway.
+
+        The gateway handler disconnects this zombie adapter and requeues the
+        platform for background reconnection (retryable) or marks it fatal
+        (non-retryable). Callers must ``return`` immediately after this.
+
+        ``_set_fatal_error`` runs before ``_mark_disconnected`` so the latter
+        sees ``has_fatal_error`` and skips its redundant status write.
+        """
+        self._set_fatal_error(code, message, retryable=retryable)
+        self._mark_disconnected()
+        self._notify_fatal_error_detached()
+
     async def _listen_loop(self) -> None:
         """Read WebSocket events and reconnect on errors.
 
@@ -565,7 +609,12 @@ class QQAdapter(BasePlatformAdapter):
                             "Check: 1) AppID/Secret correct 2) Bot permissions on QQ Open Platform",
                             self._log_tag,
                         )
-                        self._set_fatal_error(
+                        # This branch returns before the shared
+                        # _fail_pending("Connection closed") below, so fail
+                        # pending response futures here — a handler-less
+                        # (standalone) adapter would otherwise never settle them.
+                        self._fail_pending("Too many quick disconnects")
+                        self._give_up_and_notify(
                             "qq_quick_disconnect",
                             "Too many quick disconnects — check bot permissions",
                             retryable=True,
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: #72673 addresses the reconnect-exhaustion handoff. This PR uses a guarded detached notification and also covers the previously unnotified quick-disconnect and fatal-close exits; please consolidate the competing QQBot recovery approaches.

ttxs69:
Acknowledged — and thanks to @cadezhou, who independently diagnosed the same silent-death root cause a month earlier in #72673. I searched for competing work before opening but missed that PR (different vocabulary: "silent death" vs "zombie/not connected").

**Relationship:** both PRs implement the same core mechanism — detached fatal-error notification so the gateway watcher takes over, with the same deadlock rationale (gateway handler → `disconnect()` → awaits `_listen_task`). The remaining delta:

| | #72673 | This PR |
|---|---|---|
| 3× reconnect-exhaustion exits | ✅ | ✅ |
| quick-disconnect exit | ✅ (added after sweeper review) | ✅ |
| fatal close-code exits (banned/offline/…) | ❌ still `_set_fatal_error` without notify → same zombie | ✅ notified, non-retryable |
| notify-task exception guard | ❌ bare `create_task(self._notify_fatal_error())` — a raising gateway handler leaves an unawaited task error and the notification is lost | ✅ guarded best-effort wrapper |
| quick-disconnect `_fail_pending` | ❌ | ✅ (defensive parity) |
| error-code context (last close code / exc) | ✅ | ✅ absorbed in 57499c3cd8, credit in commit message |
| tests | mock `_notify_fatal_error` (noted by sweeper) | real notify chain, handler calls `disconnect()` (deadlock guard), mutation-tested |

**Suggested consolidation:** take this PR as the base — it is a strict superset after absorbing the error-code context, and #72673's remaining gaps are the two real defects above (unguarded notify task, unnotified fatal-close exits). @cadezhou's authorship of the core mechanism deserves credit and is acknowledged in the commit message and PR description.

If the team prefers #72673 as the base instead, I'm happy to port the three fixes there as commits — whichever unblocks the merge fastest works for me. The bug has been live on main for a long time (we hit it in production: a 2h DNS outage on 2026-08-19 left qqbot dead for 10 days).

ttxs69:
Update for the maintainers — full landscape of the four open PRs touching this zombie-adapter bug (consolidation requested by triage on both #72673 and this PR):

| PR | Age | Approach | Status |
|---|---|---|---|
| #19414 | May 3 | 5-bug batch (zombie + close codes + cooldown + heartbeat ×2), +750/−33 | Test quality concerns — e.g. `test_sleep_present_in_reconnect_source` asserts the *source string* contains `asyncio.sleep(15)`; several "method exists" tests. No progress in ~4 months. |
| #17814 | older | `_set_fatal_error` + **inline `await self._notify_fatal_error()`** in the listener | The inline await is the hazard both #72673 and this PR avoid: the gateway handler calls `adapter.disconnect()`, which awaits `self._listen_task` — from inside the listener that self-awaits (deadlock) / cancels the carrier mid-handler, so the platform never reaches `_failed_platforms`. |
| #72673 | Jul 27 | detached notify (same core as here) | Correct core mechanism; missing exception guard on the notify task, fatal-close-code exits still unnotified, quick-disconnect `_fail_pending` (details in my comment there). Also includes a valuable mock-free `GatewayRunner` handoff e2e test. |
| this PR | today | detached + guarded notify, all 5 exits, `_fail_pending`, slugified codes | Strict superset of #72673 after absorbing its error-code context (credit in 57499c3cd8). Two independent review rounds; tests mutation-verified. |

Suggestion to unblock: converge on one base (this PR or #72673 — I'll port my deltas to theirs if preferred), close the superseded ones with credit, and treat the orthogonal pieces separately (#19414's cooldown/heartbeat items are a different scope; #17814's approach is superseded by the detached design). Happy to do the porting/cleanup work either way — the bug has P2 labels on three of these PRs and has been live since well before May.

ttxs69:
Follow-up: the mock-free `GatewayRunner` handoff e2e test from #72673 has been adapted into this PR (18b9459902, error codes adjusted, credit in the commit message) — a retryable fatal now provably reaches the real `_handle_adapter_fatal_error` and lands in `_failed_platforms` with `attempts=0`. Combined coverage is now adapter-side (4 exit tests, mutation-verified) + runner-side e2e, strictly superseding both PRs individually. 69 passed.

PR #97858, opened 2026-08-29
feat(browser): support Dia as a real-profile Chromium browser

dappermint:
## Summary

Dia (The Browser Company) is a Chromium browser, but real-profile browsing rejected it as non-Chromium. On a machine where Dia is the OS default, `browser.use_real_profile: true` fails every time with:

> browser.use_real_profile is on, but your default browser is not a supported Chromium browser (Chrome, Edge, Brave, Chromium).

It turned out to be missing detection, not a missing capability. I verified Dia speaks CDP with nothing stripped **before** writing any code:

```bash
open -na /Applications/Dia.app --args --remote-debugging-port=9333 \
  --user-data-dir=/tmp/dia-cdp-test --no-first-run
curl -s http://127.0.0.1:9333/json/version
```
```json
{ "Browser": "Chrome/152.0.7977.65", "Protocol-Version": "1.3" }
```

Its profile is at `~/Library/Application Support/Dia/User Data` in the standard Chromium layout (`Local State` + `Default/`), so the existing snapshot path needs no changes at all.

## What changed

Dia is registered at the five sites in `browser_connect.py` that already enumerate supported browsers:

| Site | Addition |
|---|---|
| `_DARWIN_APPS` | `/Applications/Dia.app/Contents/MacOS/Dia` |
| `_CHROMIUM_BROWSERS` | `"dia"` |
| `_DARWIN_BUNDLE_MAP` | `("company.thebrowser.dia", "dia")` |
| `_real_profile_relparts` | `("Dia", "User Data")` |
| `chromium_executable` (Darwin) | Dia binary path |

The four user-facing strings that enumerate supported browsers are updated alongside (tool error messages, config comment, docs, desktop settings copy). Leaving them stale would tell users Dia is unsupported while it demonstrably works.

macOS-only, matching where Dia ships. Linux/Windows tables are untouched.

## Test plan

- [x] `tests/hermes_cli/test_browser_connect_default_chromium.py` — **31 passed**
- [x] New bundle-map case is **mutation-verified**: with the `_DARWIN_BUNDLE_MAP` entry removed, `test_bundle_map[company.thebrowser.Dia-dia]` fails while the other four still pass. It guards the mapping rather than restating it.
- [x] No regression: `chrome` / `brave` / `edge` / `chromium` profile paths and detection are byte-identical before and after.
- [x] End-to-end against real imports — `detect_default_chromium()` returns `dia`, `real_profile_data_dir("dia")` resolves to a directory that exists and contains `Default/`.

The 18 errors from the wider `-k 'browser_connect or real_profile'` selection are a missing `fastapi` in my test sandbox, and reproduce identically on unmodified `main`. Unrelated to this change.

## Note for maintainers (separate, pre-existing issue)

While testing this I hit a bug that is **not** Dia-specific and is **not** addressed here, but is worth a look:

`snapshot_real_profile()` → `_mirror_profile_auth()` hangs on `Web Data` when the source browser is **running**. `Cookies`, `Login Data` and `Login Data For Account` all mirror fine, but `Web Data` stays 0 bytes indefinitely — sqlite's online-backup restarts its pass whenever the source is written, and a live browser writes that DB continuously, so it never converges.

The fast lock probe intentionally no-ops on POSIX ("copy-while-running still works"), which holds for the other DBs but not this one. Quitting the browser first makes the snapshot complete in seconds. Happy to open a separate issue if useful.

### DIFF
--- a/apps/desktop/src/app/settings/constants.ts
+++ b/apps/desktop/src/app/settings/constants.ts
@@ -558,7 +558,7 @@ export const FIELD_DESCRIPTIONS: Record<string, string> = defineFieldCopy({
   timezone: 'IANA timezone identifier. Blank uses the system timezone.',
   browser: {
     useRealProfile:
-      "Local browsing uses your real logins. Hermes copies your default browser's profile (cookies, logins, preferences) into a managed snapshot and drives it with its packaged Chromium — your live profile is never opened directly, and the copy is refreshed from it on each run. Also lets the agent open a local real-profile session on request even when a cloud browser backend is configured. Only Chromium browsers (Chrome, Edge, Brave, Brave Origin, Chromium) are supported; a non-Chromium default fails with a clear message. Off by default."
+      "Local browsing uses your real logins. Hermes copies your default browser's profile (cookies, logins, preferences) into a managed snapshot and drives it with its packaged Chromium — your live profile is never opened directly, and the copy is refreshed from it on each run. Also lets the agent open a local real-profile session on request even when a cloud browser backend is configured. Only Chromium browsers (Chrome, Dia, Edge, Brave, Brave Origin, Chromium) are supported; a non-Chromium default fails with a clear message. Off by default."
   },
   agent: {
     imageInputMode: 'Controls how image attachments are sent to the model.',
--- a/hermes_cli/browser_connect.py
+++ b/hermes_cli/browser_connect.py
@@ -96,6 +96,11 @@ _BROWSERS = (
         ("/usr/bin/brave-origin", "/opt/brave.com/brave-origin/brave-origin",
          "/opt/brave.com/brave-origin-nightly/brave-origin"),
         "BraveSoftware/Brave-Origin", linux_exec=("brave-origin",)),
+    # Dia (The Browser Company): Chromium core, macOS-only. Windows is announced but has no
+    # public build, so the Windows/Linux fields stay empty and resolution fails closed there.
+    _Browser(
+        "dia", "/Applications/Dia.app/Contents/MacOS/Dia",
+        ("Dia", "User Data"), (), (), (), (), (), ""),
     _Browser(
         "edge", "/Applications/Microsoft Edge.app/Contents/MacOS/Microsoft Edge",
         ("Microsoft Edge",), ("msedge.exe", "msedge"),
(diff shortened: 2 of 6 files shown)


Enough1122:

**feat(browser): support Dia as a real-profile Chromium browser**

Reasonable, consistent feature addition: Dia is threaded through `_CHROMIUM_BROWSERS`, the macOS app path list, the Darwin bundle map (`company.thebrowser.dia`), `_real_profile_relparts`, config defaults, docs, user-facing error messages, and the bundle-map test.

Non-blocking observations:

1. **Platform coverage appears macOS-only.** The diff adds Dia to the Darwin app paths, Darwin bundle map, and relparts — but I don't see a corresponding entry in `_WINDOWS_BROWSER_GROUPS` (Windows install paths / `shutil.which` bin names) or a Linux group. So on Windows/Linux, `get_chrome_debug_candidates()` won't discover Dia, and real-profile driving of a Dia default browser likely won't resolve. If Dia is intended to work beyond macOS (the docs say "whichever is your OS default"), the Windows/Linux detection entries are missing; if macOS-only is the intent, consider documenting that.

2. **Identical relparts tuples.** `_real_profile_relparts` for `"dia"` is `(("Dia", "User Data"), ("Dia", "User Data"), "Dia/User Data")` — both tuple entries identical. If those entries are meant to be platform-specific (e.g., a Windows `Dia\User Data` layout differing from macOS), double-check the second entry; as-is it's harmless but looks copy-pasted.

3. **Bundle-id case sensitivity.** The test pins `company.thebrowser.Dia` (mixed case) → `"dia"`, so the matcher normalizes case — good, and worth keeping that test as the guard.

dappermint:
Rebased onto main. On the platform-coverage point: Dia has no Windows or Linux build (Windows is announced, waitlist only), so there's nothing to detect there and I'd rather not predict install paths. Instead the empty platform fields now fail closed: `real_profile_data_dir` used to return `LOCALAPPDATA` / `~/.config` themselves when `win_profile` / `linux_config` were empty, which would snapshot a parent dir. Both branches return None now, pinned by `TestMacOnlyBrowser` (mutation-checked). The duplicate relparts tuple is gone with the rebase, that table no longer exists. Docs say "Dia on macOS".

PR #97860, opened 2026-08-29
fix(kanban): never auto-reclaim inline-owned running cards

JCarri14:
Fixes the inline-owner vs dispatched-worker double-owner / nudge-loop / crash cascade.

## Root cause
An INLINE owner (CLI `kanban claim` / gateway session) holds a card `running` with `worker_pid=NULL`. The TTL (`release_stale_claims`) and heartbeat-staleness (`detect_stale_running`) reclaim passes treated it like any expired claim and requeued it to `ready`, where the dispatcher immediately spawned a duplicate worker beside the inline owner → `protocol_violation` loop → `pid not alive` crashes → `failure_limit` / `gave_up`.

## Change
- `release_stale_claims` and `detect_stale_running` now only consider `worker_pid IS NOT NULL` (dispatcher-spawned workers). Inline-owned cards are never auto-reclaimed.
- `detect_crashed_workers`/`enforce_max_runtime` already filtered `worker_pid IS NOT NULL`; `reconcile_orphaned_running` is structurally safe (a valid inline claim always carries both claim_lock and claim_expires).
- Operator `reclaim_task` remains the only recovery for a genuinely dead inline session.

## Preserves
- Dead dispatcher workers are still reclaimed → crash recovery intact (regression test).
- `HERMES_DELEGATED_CHILD_CONTEXT` guard untouched.
- ONE CARD → ONE EXECUTOR.

## Tests
- New `tests/hermes_cli/test_kanban_inline_ownership.py` (2 tests).
- 96 existing kanban DB/gateway/tools/worker-runs tests pass.

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -4478,6 +4478,23 @@ def release_stale_claims(
     API traffic. ``enforce_max_runtime`` and ``detect_crashed_workers``
     remain the upper bounds for genuinely wedged or dead workers.
 
+    Inline ownership guard: only tasks with a dispatcher-spawned worker
+    (``worker_pid`` set, i.e. status='running' with a real subprocess)
+    are reclaim candidates in the TTL and heartbeat-staleness passes. A
+    ``running`` task whose ``worker_pid`` is NULL is held by an INLINE
+    owner — a CLI ``kanban claim`` or a gateway/admin session, not the
+    dispatcher. Auto-reclaim must never release such a claim: doing so
+    requeues the card to ``ready`` and the dispatcher spawns a duplicate
+    worker beside the inline owner, producing the nudge-loop / double-
+    owner / protocol-violation crash cascade (see the hardening task on
+    inline-owner vs dispatched-worker ownership). A legitimate inline
+    claim always carries both ``claim_lock`` and ``claim_expires``, so it
+    is also structurally excluded from orphan reconciliation (which only
+    touches cards missing either field). The inline owner manages its own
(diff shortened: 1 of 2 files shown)


Enough1122:

**fix(kanban): never auto-reclaim inline-owned running cards**

Well-targeted fix with a clear ownership model. Adding `AND worker_pid IS NOT NULL` to the TTL pass in `release_stale_claims` and to `detect_stale_running` means inline owners (CLI `kanban claim` / gateway sessions, `worker_pid` NULL) are never requeued to `ready` by auto-reclaim — closing the duplicate-worker / nudge-loop cascade. Orphan reconciliation already structurally excluded inline owners (they carry `claim_lock` + `claim_expires`), so the model is internally consistent. The two tests pin both halves: inline-owned cards stay `running` (and `dispatch_once` spawns nothing), while a dead dispatcher worker is still reclaimed (crash recovery preserved).

Non-blocking observations:

1. **Stale telemetry now hides inline owners.** `detect_stale_running` also feeds `DispatchResult.stale` in `dispatch_once` (the line ~9338 call site), which is used for telemetry/monitoring. With the new filter, a genuinely hung *inline* session no longer appears in stale telemetry — it's only recoverable via operator `reclaim_task` (documented intent). If any dashboard or alert consumes `result.stale`, consider surfacing inline-owned stale cards separately (e.g., a distinct flag) so they aren't silently invisible, even though auto-reclaim correctly stays hands-off.

2. **Trivial file nit.** `tests/hermes_cli/test_kanban_inline_ownership.py` ends with "No newline at end of file". Cosmetic.

PR #97864, opened 2026-08-29
fix(kanban): complete durable orchestration lifecycle

sergei-ivanisenko:
## Summary
- enforce explicit `agent` / `external` / `no_agent` ownership and central installed-profile admission across DB, CLI, tools, dispatcher, and dashboard
- add durable task-owned external runs with exact process identity, canonical exit receipts, progress, bounded retries, crash recovery, typed waits, and automatic continuation
- bind candidates, evidence manifests, Reviewer/VisualQA gates, and verdicts to immutable exact-SHA/run provenance with coordinator-only control authority
- add a deterministic single-writer release barrier with lease/CAS, immutable delivery intent, receipt, and exact readback
- expose external lifecycle state in CLI/dashboard and preserve additive legacy migrations, including drifted-table rebuilds

## Verification
- focused orchestration/provenance/process suites: `99 passed`
- broad no-retry Kanban/process suite: `68 files, 777 passed, 0 failed, 8 platform-skipped`
- final exact-SHA independent reviews: `APPROVE` / `APPROVE`
- `ruff check .`, `py_compile`, dashboard `node --check`, and `git diff --check`: passed

Exact reviewed head: `c940a0e93bd85bfd56600b2ae8100afe8f1fa2c7`

### DIFF
--- a/agent/kanban_stop.py
+++ b/agent/kanban_stop.py
@@ -13,6 +13,7 @@ loop continues instead of exiting.
 
 from __future__ import annotations
 
+import json
 import os
 from typing import Any, Iterable, Optional
 
@@ -47,6 +48,17 @@ def _tool_call_name(tc: Any) -> str:
     return str(getattr(tc, "name", "") or "")
 
 
+def _successful_process_transfer(msg: dict) -> bool:
+    """Whether a tool result records the process tool's durable handoff receipt."""
+    if str(msg.get("name") or "") != "process":
+        return False
+    try:
+        receipt = json.loads(str(msg.get("content") or ""))
+    except (TypeError, ValueError):
+        return False
+    return isinstance(receipt, dict) and receipt.get("status") == "transferred" and bool(receipt.get("run_id"))
+
+
 def session_called_kanban_terminal(messages: Iterable[dict] | None) -> bool:
     """True if this conversation already invoked a terminal kanban tool."""
     if not messages:
@@ -60,6 +72,8 @@ def session_called_kanban_terminal(messages: Iterable[dict] | None) -> bool:
                 if _tool_call_name(tc) in _TERMINAL_KANBAN_TOOLS:
                     return True
         elif role == "tool":
+            if _successful_process_transfer(msg):
+                return True
             name = str(msg.get("name") or "")
             if name in _TERMINAL_KANBAN_TOOLS:
                 return True
--- a/hermes_cli/kanban.py
+++ b/hermes_cli/kanban.py
@@ -81,6 +81,8 @@ def _task_to_dict(t: kb.Task) -> dict[str, Any]:
         "session_id": t.session_id,
         "workflow_template_id": t.workflow_template_id,
         "current_step_key": t.current_step_key,
+        "wait_kind": t.wait_kind,
+        "wait_ref": t.wait_ref,
     }
 
 
@@ -398,6 +400,14 @@ def build_parser(parent_subparsers: argparse._SubParsersAction) -> argparse.Argu
                                "and re-queues the task.")
     p_create.add_argument("--created-by", default="user",
                           help="Author name recorded on the task (default: user)")
+    p_create.add_argument("--owner-kind", choices=("agent", "external", "no_agent"), default=None)
+    p_create.add_argument("--task-kind", default="ordinary")
+    p_create.add_argument("--purpose", default=None)
+    p_create.add_argument("--created-by-task-id", default=None)
+    p_create.add_argument("--created-by-run-id", type=int, default=None)
+    p_create.add_argument("--creation-authority", default=None)
+    p_create.add_argument("--gate-candidate-sha", default=None)
+    p_create.add_argument("--gate-manifest-hash", default=None)
     p_create.add_argument("--skill", action="append", default=[], dest="skills",
                           help="Skill to force-load into the worker "
                                "(repeatable). The kanban lifecycle is already "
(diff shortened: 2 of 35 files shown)


sergei-ivanisenko:
Independent exact-SHA review: **APPROVE** for `4cd38599031587ef8bd63debb83b1e64bb980145` with no blocking issues. Local verification: 770 passed, 0 failed, 7 platform-skipped; Ruff and `git diff --check` passed.

andrexibiza (commented):
Reviewed exact head `c940a0e93bd85bfd56600b2ae8100afe8f1fa2c7` against exact base/current `main@4209d371aa1bb8840ce8447555bdd863a1a96c38`, including the 32-file diff, process handoff/recovery, candidate + evidence provenance, reviewer/VisualQA gates, release barrier, tests, current CI, and the live Kanban PR collision set.

There is a lot of good work here. In particular, the transferred-process path now carries `host_start_time` as a PID-reuse guard, checkpoints pending settlement before relinquishing session ownership, and makes the DB external run authoritative after transfer. The candidate/gate path also does real exact-SHA/run binding rather than reconstructing provenance from mutable task state. Those are meaningful improvements.

I found two P1 authority/settlement blockers, plus one structural landing problem.

### P1 — `HERMES_PROFILE` is being used as authentication for coordinator-only mutations, but the subject can rewrite it

`hermes_cli/kanban_db.py::_require_control_authority()` derives privileged authority from `os.environ["HERMES_PROFILE"]` and compares that string to `kanban.coordinator_profile`. `hermes_cli/kanban.py::_bound_cli_control_authority()` repeats the same check. This is not an authority boundary for a Kanban worker with terminal/code execution: the worker can invoke the same public CLI as, for example,

`HERMES_PROFILE=<configured-coordinator> hermes kanban ...`

or set `os.environ` before importing the DB API. The configured coordinator name is ordinary configuration, not a secret or unforgeable runtime identity. `_assert_not_delegated_child_mutation()` only excludes delegated-child contexts; it does not make a normal implementation worker unable to rewrite its child environment.

The consequence is broader than one verb. The same check gates reviewer/VisualQA task creation, scratch/dir candidate attestation, gate-verdict recording, capability/human wait receipts, release-barrier creation/lease/delivery/readback, and reconciliation. An implementation worker can therefore promote itself to the coordinator principal at the exact boundary this PR says is coordinator-only.

The current test actually encodes the vulnerable model: `test_evidence_freeze_rejects_authority_spoofing_runtime_profile` treats `monkeypatch.setenv("HERMES_PROFILE", "coordinator")` as establishing coordinator authority. That proves string consistency, not principal ownership.

Required repair: derive coordinator authority from something the constrained child cannot mint or rewrite — e.g. a parent-owned capability/carrier/token bound to the task/run/generation and checked at the DB mutation boundary. Keep the supplied profile/name only as metadata/assertion. Add an adversarial regression that starts from a non-coordinator Kanban worker, overrides `HERMES_PROFILE`, and proves every privileged mutation remains pre-effect refused. This is the same defect class currently visible in #98127, where environment-controlled caller markers are also being treated as authorization evidence; the two fixes should converge on one authority mechanism rather than inventing two string-based gates.

### P1 — the release barrier records `delivery` and `readback` as caller assertions, so it can certify a publication that was never observed

The release state machine is internally tidy, but its external evidence boundary is missing. The CLI exposes:

- `release-barrier-delivery --target ... --sha ... --idempotency-key ...`
- `release-barrier-readback --sha ...`

and `_cmd_release_barrier()` passes those strings directly to `record_release_delivery()` / `record_release_readback()`. In the DB layer, `record_release_delivery()` stores the caller-supplied `delivered_sha`; `record_release_readback()` stores the caller-supplied `readback_sha`; then `reconcile_release_barrier()` reaches `released` when those stored strings equal the candidate SHA and the gates pass.

There is no target adapter/read API, target-issued receipt, content digest, remote object identity, or independently derived readback in that path. The same coordinator that says "I delivered SHA X" can immediately say "I read back SHA X" without any external effect. The lease/CAS makes that self-assertion exactly-once, but it does not make it true.

Required repair: make delivery settlement consume evidence produced by the actual mutation/readback boundary. The publication adapter should return a typed delivery receipt binding target identity + idempotency key + exact object/version; a separate readback operation must derive the observed object/version from the target and bind it to that receipt. `reconcile_release_barrier()` should accept only those verified receipts, not free-form SHA arguments from the same actor. Add a negative test where the publication is skipped but the caller supplies matching SHAs; the barrier must remain unreleased.

### Structural / merge-order blocker — this PR adds another 1,747 lines to `hermes_cli/kanban_db.py` and currently collides with active Kanban work

This patch adds 1,747 lines to `hermes_cli/kanban_db.py`, which is already well beyond 15k lines. The new ownership model, external-run lifecycle, wait receipts, candidate/evidence provenance, gate validation, and release-barrier protocol are separable authorities; putting all of them back into the existing DB godfile makes subsequent verification and composition materially harder. Please extract the new domains behind narrow sub-2k modules/interfaces and leave `kanban_db.py` as the compatibility/data-access facade rather than growing the monolith again.

There are also live FILE-LIST collisions that need explicit composition rather than whichever branch lands last:

- #97462 (liuhao1024) and #98153 (shellybotmoyer) both modify `hermes_cli/kanban_db.py`.
- #98064 (ajseo-gp) modifies `hermes_cli/kanban_db.py` and `tools/kanban_tools.py` and is complementary durable child-orchestration work.
- #98176 (vova-superstrata) modifies `tools/kanban_tools.py` and its tests with the idempotent block contract.
- #97993 (vladikMSU) modifies `plugins/kanban/dashboard/plugin_api.py` and its tests with the bounded owner-read contract.

These are complementary/adjacent, not reasons to discard this PR. Preserve their contributor lineages and compose the semantics deliberately, especially #98153's four-eyes guard with this PR's new frozen candidate/gate machinery.

### Exact-object status

The branch is exactly based on current `main@4209d371aa1bb8840ce8447555bdd863a1a96c38`, so there is no current base-drift ambiguity. The local `99 passed` / `777 passed` evidence is useful development evidence, but the exact-head GitHub runs are not green: CI `33279406180`, Docker `33279405838`, and Nix `33279405837` are all `action_required` with no executed jobs. The only public PR review-style comment I found is for older head `4cd38599031587ef8bd63debb83b1e64bb980145`; there were no formal submitted reviews on this current head before this one.

The process-transfer/recovery work is strong and worth preserving. The remaining issues are concentrated at the two places where this PR claims the strongest guarantees: who is allowed to mint control-plane truth, and what evidence is allowed to settle an external publication. Fix those boundaries, extract the new authorities out of the godfile, compose the live Kanban collisions, then rerun exact-head CI. 🚀

sergei-ivanisenko:
Final exact candidate: `ddc18565b56d71e070894f73f83bee930c19beaa`

Release verification:
- Changed orchestration suite: **476 passed, 7 skipped** (`pytest-asyncio` + ACP extras enabled)
- Ruff, `py_compile`, Node syntax, and `git diff --check`: pass
- Independent exact-SHA Reviewer: **APPROVE**
- Independent exact-SHA compatibility review: **APPROVE**

The final remediation also preserves lossless signal exit codes, UTF-8 ConPTY input, POSIX PTY geometry/resize and single-echo behavior, and immediate parent-death durability.

PR #97871, opened 2026-08-29
fix(desktop): surface Tailscale SSH browser checks

### LINKED ISSUE #97867: [Bug]: Desktop hides Tailscale SSH browser checks behind a generic timeout
## Bug Description

Hermes Desktop hides a Tailscale SSH interactive browser check behind a generic SSH timeout.

Desktop correctly launches SSH non-interactively with `BatchMode=yes`. When a tailnet SSH policy requires a fresh browser check, Tailscale writes an actionable challenge to stderr and waits:

```text
# Tailscale SSH requires an additional check.
# To authenticate, visit: [ONE-TIME URL REDACTED]
```

After the connection timeout, Desktop reports only that the SSH operation timed out. The user cannot tell this apart from a half-open connection, slow host, firewall, or dead SSH service.

## Steps to Reproduce

1. Configure Tailscale SSH with a `check` rule whose check period has expired.
2. Confirm that an interactive `ssh user@host true` prints the Tailscale browser-check prompt and waits for the check to complete.
3. Add the same host through Hermes Desktop's **Connect via SSH** mode.
4. Test or select the connection without first completing the check in Terminal.
5. Wait for Desktop's SSH connection timeout.

## Expected Behavior

Desktop should identify the Tailscale interactive check and explain that managed SSH is non-interactive. It should tell the user to run `ssh user@host true` in Terminal, complete the browser check, and retry.

The one-time Tailscale authentication URL must not be copied into Desktop errors or lifecycle logs.

## Actual Behavior

Desktop reports the generic timeout message:

```text
SSH operation to user@host timed out. The connection may be half-open (e.g. after sleep); reconnecting.
```

## Affected Component

Hermes Desktop, managed SSH transport.

## Environment

- Client: Hermes Desktop on macOS, Apple Silicon
- Desktop/backend version observed: v0.20.6 (+1)
- Source reproduction: current `main` at `9d9f44d63826b18503f44c754e48e1f4f83b3a6e`
- Remote: Linux over Tailscale SSH

## Debug Report

Not attached because the raw Desktop and SSH logs contain private host, path, and one-time authentication data. The exact relevant stderr is sanitized above. The failure was also reproduced in a focused `ssh-connection` regression test.

## Root Cause Analysis

`apps/desktop/electron/ssh-connection.ts` accumulates child stderr inside `runSsh()`, but the timeout path rejects with a new `SSH_ERROR.TIMEOUT` error that does not preserve even a classified form of that stderr. `SshConnection._fail()` then immediately converts every timeout to `SSH_ERROR.TIMEOUT`, so `classifySshError()` never sees the Tailscale challenge.

This is distinct from:

- #96742 / #96914, which cover v1-to-v2 SSH registry migration;
- #88994 and its open fixes, which cover local-to-remote profile-name mapping;
- #96047, which evicts a wedged ControlMaster after an exec timeout;
- #97392, which extends backend-ready timeouts for slow hosts.

Searches across open, closed, and merged issues/PRs found no existing report for preserving and safely classifying the Tailscale browser-check challenge.

## Proposed Fix

1. Add a dedicated SSH error classification for a Tailscale interactive browser check.
2. On timeout, preserve only the classified stderr kind on the internal timeout error, not the raw one-time URL.
3. Let `_fail()` prefer that safe interactive-auth classification over the generic timeout.
4. Return actionable Terminal pre-authentication guidance without exposing the challenge URL.
5. Add a regression test where the SSH child emits the Tailscale challenge and remains open until the hard timeout.

## PR

I have a focused patch and regression test in progress and will link the PR here.

arcabotai:
## What does this PR do?

Hermes Desktop now recognizes when Tailscale SSH is waiting for an interactive browser check instead of reporting the wait as a generic half-open connection timeout.

The timeout path keeps only a safe classification of buffered SSH stderr. It does not retain or surface Tailscale's one-time authentication URL. Desktop tells the user to complete the check from Terminal and retry.

This is deliberately narrower than changing timeout lengths or SSH lifecycle behavior. Generic timeouts still use the existing half-open connection message.

## Related Issue

Fixes #97867

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `apps/desktop/electron/ssh-connection.ts`
  - classifies Tailscale SSH browser-check output as interactive authentication;
  - carries only that classification across the hard-timeout boundary;
  - returns safe, actionable Terminal pre-authentication guidance.
- `apps/desktop/electron/ssh-connection.test.ts`
  - lets hanging fake SSH children emit stderr before timeout;
  - reproduces the Tailscale challenge and verifies the one-time URL never reaches the user-facing error.

## How to Test

1. Run `cd apps/desktop && npx vitest run --project electron electron/ssh-connection.test.ts`.
2. Confirm the Tailscale regression reports `interactive-auth`, includes `ssh user@host true` guidance, and excludes `login.tailscale.com` plus the one-time URL token.
3. Run `npx prettier --check electron/ssh-connection.ts electron/ssh-connection.test.ts`, ESLint on both files, `npx tsc -p tsconfig.electron.json --noEmit`, and the full Electron Vitest project.

### Code

- [ ] I've run `pytest tests/ -q` and all tests pass - N/A, this is isolated TypeScript/Electron code
- [x] I've tested on my platform: Linux contract tests; the original failure was observed on macOS Apple Silicon

- [x] I've updated relevant documentation (README, `docs/`, docstrings) - N/A, no public configuration or workflow changed

## Verification

Blaxel Linux microVM, Node `v22.22.1`, npm `10.9.8`:

- RED: focused suite reproduced the bug with exactly one expected failure (`timeout` instead of `interactive-auth`).
- GREEN: focused Desktop SSH suite **68 passed**.
- `prettier --check`: passed.
- ESLint on the two changed files with `--max-warnings=0`: passed.
- Electron TypeScript typecheck: passed.
- An exploratory broader Electron run passed 1,929 tests and skipped 6. Four tests and two suites could not run because the slim worker intentionally omitted Python and Electron's postinstall binary. Those failures were outside `ssh-connection` and were classified as worker-prerequisite failures, not patch regressions.
- Live patched macOS artifact was not rebuilt. The real failure was observed there; the fix is verified at the process/error-contract boundary.

## Provenance

Prepared by Cad, an AI agent, under Luis Felipe Abarca's direction. The `arcabotai` contributor identity is human-stewarded by Luis Felipe, who authorized this issue and pull-request submission. Public background and project context: [felirami.com](https://felirami.com), [arcabot.ai](https://arcabot.ai), and [arca.computer](https://arca.computer).

### DIFF
--- a/apps/desktop/electron/ssh-connection.test.ts
+++ b/apps/desktop/electron/ssh-connection.test.ts
@@ -221,6 +221,16 @@ function fakeChild({ code = 0, stdout = '', stderr = '', errorEvent = null, hang
   }
 
   if (hang) {
+    process.nextTick(() => {
+      if (stdout) {
+        child.stdout.emit('data', Buffer.from(stdout))
+      }
+
+      if (stderr) {
+        child.stderr.emit('data', Buffer.from(stderr))
+      }
+    })
+
     return child // never emits close → drives the timeout path
   }
 
@@ -413,6 +423,35 @@ test('open() surfaces a classified auth error', async () => {
     (err: any) => {
       assert.equal(err.kind, SSH_ERROR.AUTH_FAILED)
       assert.match(err.message, /ssh-agent|ssh-add/)
+
+      return true
+    }
+  )
+})
+
+test('open() turns a Tailscale browser-check timeout into safe interactive-auth guidance', async () => {
+  const checkUrl = 'https://login.tailscale.com/a/example-one-time-check'
+
+  const spawnFn = scriptedSpawn(args => {
+    if (args.includes('check')) {
+      return { code: 255 }
+    }
+
+    return {
+      hang: true,
+      stderr: `# Tailscale SSH requires an additional check.\n# To authenticate, visit: ${checkUrl}\n`
+    }
+  })
+
+  const conn = new SshConnection({ host: 'box', user: 'me' }, { spawnFn, controlDir: '/tmp/d', connectTimeoutMs: 20 })
+
+  await assert.rejects(
+    () => conn.open(),
+    (err: any) => {
+      assert.equal(err.kind, SSH_ERROR.INTERACTIVE_AUTH)
+      assert.match(err.message, /Tailscale SSH requires an interactive browser check/)
+      assert.match(err.message, /ssh me@box true/)
+      assert.doesNotMatch(err.message, /login\.tailscale\.com|example-one-time-check/)
 
       return true
     }
--- a/apps/desktop/electron/ssh-connection.ts
+++ b/apps/desktop/electron/ssh-connection.ts
@@ -324,6 +324,7 @@ function forwardSpec(localPort, remotePort, remoteHost = '127.0.0.1') {
 const SSH_ERROR = {
   UNREACHABLE: 'unreachable',
   AUTH_FAILED: 'auth-failed',
+  INTERACTIVE_AUTH: 'interactive-auth',
   HOST_KEY_CHANGED: 'host-key-changed',
   TIMEOUT: 'timeout',
   UNKNOWN: 'unknown'
@@ -340,6 +341,12 @@ function classifySshError(stderr) {
     )
   ) {
     return SSH_ERROR.HOST_KEY_CHANGED
+  }
+
+  if (
+    /Tailscale SSH requires an additional check|To authenticate, visit:\s*https:\/\/login\.tailscale\.com\//i.test(text)
+  ) {
+    return SSH_ERROR.INTERACTIVE_AUTH
   }
 
   if (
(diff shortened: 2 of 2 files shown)


Enough1122:

**fix(desktop): surface Tailscale SSH browser checks**

Good, security-conscious change. Detecting the Tailscale "additional check" stderr and mapping it to a new `INTERACTIVE_AUTH` error kind, then emitting non-interactive guidance (`ssh <host> true`, complete the browser check, retry) is the right UX. The one-time URL is deliberately kept out of the surfaced message (the test asserts `doesNotMatch` against `login.tailscale.com`), and `stderrKind` carries only the classification through the timeout path — properly defensive about leaking a one-time auth URL through Desktop errors or lifecycle logs.

Non-blocking observations:

1. **Classification is only consulted on the timeout path.** The `stderrKind` → `INTERACTIVE_AUTH` mapping is applied inside the `if (stderrOrErr.kind === SSH_ERROR.TIMEOUT)` branch. If `ssh` exits *quickly* (non-zero, e.g. code 255) carrying the same Tailscale "additional check" stderr rather than hanging until the timeout, the guidance won't surface. Consider classifying the immediate-exit path too, or adding a test for a fast exit with that stderr.

2. **Regex requires the https scheme and exact host.** `/To authenticate, visit:\s*https:\/\/login\.tailscale\.com\//` won't match if Tailscale ever varies the scheme or host. Low risk (the message is currently fixed), but a looser match (e.g., the "requires an additional check" line alone) would be more robust.

3. **`classifySshError` now runs on every timeout.** Setting `err.stderrKind = classifySshError(stderr)` on every timeout is negligible cost, but worth noting it's a regex pass added to an already-error path. Fine.

PR #97876, opened 2026-08-29
fix(dashboard): show the slot model as current in aux/MoA pickers

xxxigm:
## Summary

- Opening **Set Auxiliary: Vision** showed `current: glm-5.3` (the main chat model) even after Vision was already pinned to Qwen. The save itself worked; the picker header/highlight always came from `/api/model/options`.
- Auxiliary and MoA pickers now pass the slot assignment, so `current:` matches the task list and the picker opens on that provider.

## Test plan

- [x] `npx vitest run src/lib/model-picker-current.test.ts` (from `web/`, 10 passed)
- [ ] Dashboard → Models → Auxiliary Tasks → pin Vision to a model that is not the main chat model
- [ ] Click Change on Vision: header shows that Vision model/provider, not the main chat model
- [ ] Tasks still on `auto` show `current: auto (use main model)`
- [ ] MoA Change on a reference/aggregator slot shows that slot, not the main model

### DIFF
--- a/ui-tui/src/__tests__/appChromeBlockedTimers.test.tsx
+++ b/ui-tui/src/__tests__/appChromeBlockedTimers.test.tsx
@@ -295,12 +295,16 @@ describe('status-chrome timers under an occluding overlay', () => {
     nowSpy.mockReturnValue(T0 + 300_000)
     rule.clear()
     resetOverlayState()
-    await flush()
+
+    // Event-based: a single 20ms flush is not enough under CI load (the
+    // reveal frame never lands, so stdout stays empty after `clear()`).
+    await vi.waitFor(() => {
+      expect(rule.output()).toContain('6m 0s')
+    })
 
     const resumed = rule.output()
 
     // Caught up to real elapsed time, not stuck on the pre-overlay values.
-    expect(resumed).toContain('6m 0s')
     expect(resumed).toContain('✓ 5m 5s')
     expect(resumed).not.toContain('1m 0s')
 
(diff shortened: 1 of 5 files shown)


Enough1122:

**fix(dashboard): show the slot model as current in aux/MoA pickers**

Clean fix. Extracting the "what is current" logic into pure helpers (`model-picker-current.ts`) with dedicated unit tests is the right shape — the screenshot bug (aux picker showing the main chat model as `current`) is properly pinned by `test_resolvePickerCurrent_does_not_let_main_model_leak`. `currentAssignment` threading from `ModelsPage` (aux task / MoA preset) through `ModelPickerDialog` is minimal and consistent.

Non-blocking observations:

1. **Inconsistent trimming across helpers.** `assignmentToPickerCurrent` trims provider/model; `resolvePickerCurrent` returns the override un-trimmed; `formatPickerCurrentLabel` and `isAutoPickerCurrent` trim at render time. The rendered output is trimmed, so it's cosmetic — but normalizing in one place (ideally `resolvePickerCurrent`) would make the data contract more predictable.

2. **Pinned-provider-missing-from-catalog divergence.** When an aux slot is pinned to a provider that isn't in the catalog, `resolveInitialProviderSlug` falls back to the main provider for the initial highlight, but the header (`formatPickerCurrentLabel`) still shows the pinned provider/model. The highlight and header can disagree in that edge case. Acceptable, but worth a conscious look.

3. **`isCurrentProvider` passes `model: ""`.** `ProviderColumn` computes `isAutoPickerCurrent({ model: "", provider: currentProviderSlug })` — the helper ignores `model`, so it works, but the empty `model` field is slightly confusing. Trivial.

PR #97877, opened 2026-08-29
fix(desktop): exempt the newest turn from the transcript render budget

feliche93:
## Problem

On long agent sessions the transcript flickers / the scroll jumps repeatedly while a turn streams, and again in a burst when it completes.

`firstVisibleGroupIndex` (the DOM history budget in `apps/desktop/src/components/assistant-ui/thread/list.tsx`) walks turns newest-first, summing render weights until the budget is met, **including the newest turn's weight**. While an agent turn streams, its weight climbs on every store flush — and lands in bulk at completion — so the hidden-history boundary advances mid-turn. Each step drops a settled turn off the top and shrinks `scrollHeight` under the stick-to-bottom lock (overflow-anchor is disabled), which reads as flicker/jump. With multiple mounted panes the per-pane budget floor is `RENDER_BUDGET/4`, so a single tool-heavy turn crosses it many times.

The live-tail parts budget (975f4ef38d) fixed the *virtualization* tail sizing but left this *hidden-history* cut mobile.

## Fix

The newest turn spends none of the budget. The cut becomes a function of settled history only: it moves exactly when the next turn begins (its predecessor's now-final weight enters the walk once, at a natural content-change moment), and never while the current turn grows or finishes.

## Tests

- Updated the two fixtures whose expectations encode the old semantics (the crossing-turn-stays-visible intent is preserved with a newest group present).
- Added two regressions: the cut is invariant to the newest turn's weight (streaming → completed), and it advances only when the finished turn joins history.
- `npx vitest run src/components/assistant-ui/thread/list.test.ts` — 30/30 pass.

Running this patch on a source build (macOS arm64) eliminates the mid-turn flicker on long tool-heavy sessions.

### DIFF
--- a/apps/desktop/src/components/assistant-ui/thread/list.test.ts
+++ b/apps/desktop/src/components/assistant-ui/thread/list.test.ts
@@ -215,17 +215,39 @@ describe('firstVisibleGroupIndex', () => {
   })
 
   it('walks newest-first and hides everything before the turn that meets the budget', () => {
-    const groups = [group('old', 50), group('mid', 30), group('new', 30)]
+    const groups = [group('oldest', 50), group('old', 30), group('mid', 30), group('new', 999)]
 
-    // newest-first: 30 (new) < 60, +30 (mid) = 60 >= 60 → mid is the first
-    // visible group, old is hidden.
+    // The newest turn is exempt, then newest-first over history: 30 (mid) < 60,
+    // +30 (old) = 60 >= 60 → old is the first visible group, oldest is hidden.
     expect(firstVisibleGroupIndex(groups, 60)).toBe(1)
   })
 
   it('keeps whole turns intact — the turn that crosses the budget stays visible', () => {
-    const groups = [group('old', 5), group('huge', 500)]
+    const groups = [group('old', 5), group('huge', 500), group('new', 10)]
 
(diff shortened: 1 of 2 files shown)


Enough1122:

**fix(desktop): exempt the newest turn from the transcript render budget**

Well-reasoned fix with tight tests. Exempting the newest turn from the render budget directly addresses the flicker/jump: the streaming turn's weight climbs per flush and lands in bulk on completion, and counting it was walking the cut forward mid-turn, dropping settled turns and shifting `scrollHeight` under stick-to-bottom. Making the cut a pure function of settled history (moving only when the next turn begins) is a sound invariant, and the four new tests cover streaming vs completed, the next-turn advance, and the empty list.

Non-blocking observations:

1. **The budget can no longer trim a single giant newest turn.** Since the newest group is unconditionally exempt, a single very long completed turn is always rendered in full regardless of weight. The docstring declares this intentional ("the cut is a function of settled history only"), so this is a note, not an objection — just confirm it matches product expectations for pathological single-turn transcripts.

2. **`minVisible` now applies to history only.** With the newest turn inherently visible, the `minVisible` parameter governs the settled history. No caller in this diff appears to rely on `minVisible` bounding the newest turn, but worth a glance if one exists elsewhere.

PR #79124, opened 2026-08-05
refactor(web): extract gateway-topology readout into web_gateway_topology (web_server.py god-file slice R2-B)

andrexibiza:
## Summary

Web_server.py god-file slice **R2-B**: extract the gateway-topology readout from `hermes_cli/web_server.py` into `hermes_cli/web_gateway_topology.py` (plain module, no routes). Part of the repo-wide large-file decomposition (tracker #78647).

## What changed and why

- Moved the 10 gateway-topology names (port-binding constants, dead-states, TTL cache, collector + cached wrapper) byte-verbatim (window 2812–2997) into `hermes_cli/web_gateway_topology.py`
- `web_server.py` re-exports all 10 (seam identity verified `is`-identical)
- **Fix (blind-review C1):** `_collect_profile_gateway_topology_cached` now resolves the collector through the `web_deps.late()` seam, so `monkeypatch.setattr(web_server, "_collect_profile_gateway_topology", ...)` stays authoritative — the 2 pre-existing tests that were RED at the extraction commit are GREEN at the fix commit
- Repointed the cache-test monkeypatch targets to the new module
- Added seam-identity regression (10 names) + aggressive tests (cache hit/miss, TTL expiry, lock, dead-state classification)

## Why this matters to users

The /api/status topology cache exists to stop the desktop's 1/s status polls from GIL-starving the event loop (#60800). Extracting it into a focused module preserves that behavior while making the large-file decomposition possible — and the late() seam means test/hot-reload monkeypatches keep working exactly as before.

## Testing

- `pytest tests/test_web_server_gateway_topology_seam.py tests/test_web_server_status_topology_cache.py tests/hermes_cli/test_web_server_gateway_topology.py` → **31 passed**
- Canonical runner: **2 files, 24 passed, 0 failed**
- RED→GREEN proven: the 2 C1 tests fail at `caf87913cc`, pass at `0132d410f6` (the fix)
- ruff clean · git diff --check clean · LF-only · DCO signed
- Platforms tested: Windows (native), CI-parity runner

## Coordination / interlock

- **Epic:** Part of #78647 (Shard all 20 god files)
- **Related PRs:** sibling web_server slices (model-assignment, managed-files, schema, custom-endpoints, messaging, whatsapp) — disjoint clusters
- **Dedup:** no existing PR extracts the topology readout
[...]

PR #97878, opened 2026-08-29
fix: proxy platform webhooks past the dashboard OAuth gate

toyama2026:
## What does this PR do?

On any deployment where the dashboard binds a public/non-loopback host and only ONE port is publicly reachable (e.g. a Fly.io app that declares a single `[http_service]`), a platform adapter's own webhook server never gets exposed. `LINE`'s adapter, for example, runs a separate `aiohttp` server on port 8646 (`plugins/platforms/line/adapter.py`) — a real request to `/line/webhook` on the public hostname never reaches it. Instead it hits this app's global dashboard-auth middleware, which has no route for that path and, under the OAuth gate, bounces every request to `/login`. Confirmed live against a real deployment: LINE's own webhook-verification API call got a `302` (expected `200`), meaning real inbound events are silently dropped — no error surfaces anywhere, the webhook just doesn't fire.

`hermes_cli/dashboard_auth/public_paths.py` already documents the right precedent for this exact class of problem: `/api/cron/fire` bypasses the cookie gate because it carries its own short-lived JWT, which is the real auth boundary — the allowlist isn't the security control there, the JWT is. LINE, Feishu, and WeCom's callback endpoint already verify every request with their own signature/HMAC check inside the adapter (`verify_line_signature`, etc.), so routing them through the dashboard's browser-session gate is both wrong (no browser session exists for a server-to-server LINE callback) and redundant.

## Related Issue

No existing issue — opening this PR directly since the root cause and fix were both identified and verified together. Happy to file a companion issue if preferred.

## Type of Change

## Changes Made

- `hermes_cli/web_server.py`:
  - Added `_WEBHOOK_PROXY_DEFAULT_PATHS`, mapping platform name to its adapter's default webhook path (`line`, `feishu`, `wecom_callback` — see Scope below).
  - Added `_resolve_webhook_proxy_port()`, reusing the existing `_PORT_BINDING_PLATFORM_PORTS` table and the same config-precedence logic `_profile_platform_ports()` already uses for the dashboard's topology display, to find the adapter's actual live port (config override, falling back to the documented default).
  - Added `_proxy_platform_webhook()`, an `httpx.AsyncClient`-based reverse proxy that forwards method/body/headers (minus hop-by-hop headers) to `http://127.0.0.1:<port><path>` and streams the response back unmodified.
  - Added `_webhook_proxy_gate`, a new `@app.middleware("http")` registered so it intercepts a matching path before `_dashboard_auth_gate` ever sees it. Unmatched paths fall through unchanged to the existing middleware chain.
- `tests/hermes_cli/test_webhook_proxy_gate.py` (new): full coverage of the new gate — bypass + proxy for each listed platform, query-string forwarding, header stripping, unreachable-adapter to 502, config-port resolution + its fallback-on-error path, and a regression guard that an unrelated path still hits the OAuth gate exactly as before.

## Scope (intentionally conservative)

Only `line`, `feishu`, and `wecom_callback` are in the initial table — each verified directly against its adapter's source for the exact default path (`plugins/platforms/line/adapter.py`, `plugins/platforms/feishu/adapter.py`, `plugins/platforms/wecom/callback_adapter.py`). The other `_PORT_BINDING_PLATFORM_PORTS` entries (`msgraph_webhook`, `bluebubbles`, `sms`, `whatsapp_cloud`) are deliberately left out: I could not confirm their exact default webhook path against source with confidence in this pass (e.g. `sms`'s path lives under the `/webhooks/` namespace shared with the generic user-defined-webhooks feature, which needs a closer look to avoid a path collision), and a wrong guess here would silently misroute traffic rather than fail loudly. Extending the table for a verified platform is a one-line addition to `_WEBHOOK_PROXY_DEFAULT_PATHS` (plus `_PORT_BINDING_PLATFORM_PORTS` if it's not already there).

An operator-customized `extra.webhook_path` / `extra.path` (away from the adapter's default) also isn't covered — the path-match is against the *default* path only. This is no worse than today (such a setup already required fronting the platform's port directly, e.g. via its own Cloudflare Tunnel), just not yet improved by this PR.

## How to Test

1. `pytest tests/hermes_cli/test_webhook_proxy_gate.py -q` — new tests, fully mocked (no real listener required).
2. Reproduce the bug on `main` first: `TestClient` a gated `app` (see the `gated_app` fixture in `tests/hermes_cli/test_dashboard_auth_middleware.py`) and `POST /line/webhook` -> `302` to `/login`.
3. With this branch: the same request instead reaches `_webhook_proxy_gate` and proxies through (mocked in the test suite; on a machine with a real LINE adapter running locally on 8646, it forwards to the real adapter — I confirmed this by hand against my own running instance and got back a genuine `401 invalid signature` from the adapter itself, proving the proxy actually reaches the real service end-to-end).

### Code

- [x] My commit message follows Conventional Commits (`fix:`)
- [x] I searched existing PRs/issues for a duplicate report of this — didn't find one
- [x] This PR contains only changes related to this fix (no unrelated commits — my local checkout had unrelated pre-existing modifications to other files; none of them are in this branch/commit)
- [ ] I've run `pytest tests/ -q` and all tests pass — I ran the directly-affected suites instead (`tests/hermes_cli/*dashboard_auth*`, `*webhook*`, `test_web_server*.py` — 397 tests total across both runs), all passing except one pre-existing, unrelated failure (`test_custom_endpoint_save_scopes_to_the_requested_profile`, a Windows/cp932 locale `UnicodeDecodeError` in the custom-endpoint-save code path) that reproduces identically on `main` without this change. I did not run the full multi-thousand-test suite end-to-end due to time constraints — happy to let CI cover the remainder.

- [ ] I've updated relevant documentation — not yet; happy to add a short note to `website/docs/user-guide/messaging/webhooks/` (or wherever maintainers prefer) if this approach looks right
- [ ] I've updated `cli-config.yaml.example` — N/A, no new config keys (this reuses existing platform config already documented there)
- [ ] I've updated `CONTRIBUTING.md` / `AGENTS.md` — N/A, no workflow/architecture change beyond this one middleware
- [x] I've considered cross-platform impact — the proxy is a plain loopback HTTP call (`httpx.AsyncClient` to `127.0.0.1`), no platform-specific behavior
- [x] I've updated tool descriptions/schemas if I changed tool behavior — N/A, no tool-facing change

## Screenshots / Logs

Live repro (before this fix), against a real Fly.io-hosted Cloud Agent instance:

```
$ curl -s -X POST https://api.line.me/v2/bot/channel/webhook/test -H "Authorization: Bearer $LINE_CHANNEL_ACCESS_TOKEN"
{"success":false,"timestamp":"2026-08-29T11:57:02.003289Z","statusCode":302,"reason":"ERROR_STATUS_CODE","detail":"302"}

$ curl -s -o /dev/null -D - -X POST "https://<redacted>.agents.nousresearch.com/line/webhook" -d '{}'
HTTP/1.1 302 Found
location: /auth/login?provider=nous&next=%2Fline%2Fwebhook
```

`/api/health` and `/api/status` on the same host correctly return `200` (they're in `PUBLIC_API_PATHS`), confirming this is the dashboard's own auth gate intercepting an unrouted path, not an infrastructure-level block.

### DIFF
--- a/hermes_cli/web_server.py
+++ b/hermes_cli/web_server.py
@@ -3286,6 +3286,126 @@ _PORT_BINDING_PLATFORM_PORTS: Dict[str, Tuple[str, int]] = {
 # Platform states that mean the adapter is NOT serving its port right now.
 _PLATFORM_DEAD_STATES = frozenset({"fatal", "disconnected", "stopped"})
 
+# Default webhook paths for platforms whose adapter runs its own webhook
+# server on a separate port (see ``_PORT_BINDING_PLATFORM_PORTS`` above).
+# These requests never hit a real route on THIS app — only the dashboard's
+# cookie gate does, because it runs as global middleware ahead of routing.
+# Without this table, every non-loopback deployment that exposes only this
+# app's port (the common case on single-public-port PaaS hosts) silently
+# drops inbound platform webhooks: LINE/Feishu/WeCom already verify the
+# request themselves (HMAC/signature — see each adapter's own ``verify_*``
+# function), so bouncing them through the dashboard's session-cookie gate is
+# both wrong (no browser session exists to check) and redundant (the gate
+# would just be duplicating the adapter's own auth). Mirrors the existing
+# ``/api/cron/fire`` precedent in ``dashboard_auth/public_paths.py``, which
+# bypasses the same gate for the same reason — its own short-lived JWT is
+# the real auth boundary, not this allowlist.
+#
+# A path is proxied ONLY when it matches a platform's DEFAULT path exactly
+# (or a sub-path of it, e.g. LINE's ``/line/webhook/health``). An operator
+# who set a custom ``extra.webhook_path`` / ``extra.path`` away from the
+# default isn't covered — same as before this table existed. Only entries
+# verified against the adapter source are listed; extend this table (and
+# reuse ``_PORT_BINDING_PLATFORM_PORTS`` for the port) as further platforms
+# are confirmed.
+_WEBHOOK_PROXY_DEFAULT_PATHS: Dict[str, str] = {
+    "line": "/line/webhook",
+    "feishu": "/feishu/webhook",
+    "wecom_callback": "/wecom/callback",
+}
+
+
+def _resolve_webhook_proxy_port(platform: str) -> int:
+    """Resolve the live host port for ``platform``'s webhook server.
+
+    Reads this process's own config (top-level ``platforms:`` wins over
+    ``gateway.platforms:``, matching ``load_gateway_config()`` precedence),
+    falling back to the adapter's documented default port. Best-effort: any
+    read/parse failure falls back to the default rather than raising, so a
+    config hiccup degrades to "try the default port" instead of dropping
+    the webhook outright.
+    """
+    port_key, default_port = _PORT_BINDING_PLATFORM_PORTS[platform]
+    try:
+        cfg = load_config() or {}
+        gateway_cfg = cfg.get("gateway") if isinstance(cfg.get("gateway"), dict) else {}
+        block: Dict[str, Any] = {}
+        for src in ((gateway_cfg or {}).get("platforms"), cfg.get("platforms")):
+            if isinstance(src, dict) and isinstance(src.get(platform), dict):
+                block.update(src[platform])
+        extra = block.get("extra") if isinstance(block.get("extra"), dict) else {}
+        raw = block.get(port_key, (extra or {}).get(port_key, default_port))
+        return int(raw)
+    except Exception:
+        return default_port
+
+
+async def _proxy_platform_webhook(request: Request, platform: str) -> "Response":
+    """Forward ``request`` verbatim to ``platform``'s own webhook server.
+
+    The platform adapter authenticates the request itself (signature/HMAC);
+    this proxy adds no auth of its own and must only ever be reached for a
+    path matched against ``_WEBHOOK_PROXY_DEFAULT_PATHS``.
+    """
+    import httpx
+
+    port = _resolve_webhook_proxy_port(platform)
+    body = await request.body()
+    # Hop-by-hop headers must not be forwarded (RFC 7230 SS6.1); Host/
+    # Content-Length are recomputed by httpx for the new target/body.
+    excluded = {
+        "host", "content-length", "connection", "keep-alive",
+        "transfer-encoding", "upgrade", "proxy-authenticate",
+        "proxy-authorization", "te", "trailer",
+    }
+    headers = {k: v for k, v in request.headers.items() if k.lower() not in excluded}
+    target = f"http://127.0.0.1:{port}{request.url.path}"
+    if request.url.query:
+        target = f"{target}?{request.url.query}"
+    try:
+        async with httpx.AsyncClient(timeout=httpx.Timeout(30.0)) as client:
+            upstream = await client.request(
+                request.method, target, content=body, headers=headers,
+            )
+    except httpx.RequestError as exc:
+        _log.warning(
+            "Webhook proxy: %s adapter unreachable on port %s: %s",
+            platform, port, exc,
+        )
+        return JSONResponse(
+            status_code=502,
+            content={"detail": f"{platform} webhook adapter unreachable"},
+        )
+    response_headers = {
+        k: v for k, v in upstream.headers.items() if k.lower() not in excluded
+    }
+    return Response(
+        content=upstream.content,
+        status_code=upstream.status_code,
+        headers=response_headers,
+    )
+
+
+@app.middleware("http")
+async def _webhook_proxy_gate(request: Request, call_next):
+    """Forward known platform-webhook paths straight to their own server.
+
+    Must run before ``_dashboard_auth_gate``: these paths carry no browser
+    session, are authenticated by the platform's own signature check inside
+    the adapter, and — critically — don't correspond to any real route on
+    THIS app at all when the platform's adapter runs on a separate port (the
+    common single-public-port PaaS deployment, e.g. a Fly.io app that only
+    declares one public service). Letting the auth gate see them first means
+    an unauthenticated caller (LINE's own servers, with a valid signature)
+    gets bounced to /login and the webhook is silently dropped — the bug
+    this closes.
+    """
+    path = request.url.path
+    for platform, default_path in _WEBHOOK_PROXY_DEFAULT_PATHS.items():
+        if path == default_path or path.startswith(default_path + "/"):
+            return await _proxy_platform_webhook(request, platform)
+    return await call_next(request)
+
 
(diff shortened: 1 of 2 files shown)


andrexibiza (commented):
This is a real production failure, and the live `302 -> /auth/login` reproduction is useful evidence. The narrow path table is also the right instinct: refusing to guess routes for SMS/WhatsApp/etc. is much safer than turning a routing repair into an ambient catch-all. I checked the exact head `a837c3cad8f1107b5dd7cd26f863357ed486385d`, the full two-file diff, the current LINE/Feishu/WeCom callback implementations, middleware ordering, current-main ancestry, hosted checks, existing PR comments/reviews, and the open `web_server.py` decomposition graph.

I do not think this head is merge-ready yet. There are three correctness/security blockers plus one architecture/release gate.

### P1 — the proxy removes the adapters' pre-auth body-size boundary

`hermes_cli/web_server.py::_proxy_platform_webhook()` does:

```py
body = await request.body()
```

before the request reaches the adapter. There is no outer size cap or bounded stream read.

That is a semantic regression at the new public ingress boundary. The three adapters deliberately reject oversized *unauthenticated* bodies before doing expensive/authenticated work:

- LINE: `WEBHOOK_BODY_MAX_BYTES = 1_048_576`, enforced by `web.Application(client_max_size=WEBHOOK_BODY_MAX_BYTES)`.
- Feishu: `_FEISHU_WEBHOOK_MAX_BODY_BYTES = 1 * 1024 * 1024`, with both a bounded reader and `client_max_size` backstop.
- WeCom callback: `_MAX_BODY = 65_536`; its source explicitly says `client_max_size` must reject before the handler/signature work.

With this proxy in front, an unauthenticated Internet client can make the dashboard process buffer an arbitrarily large body *before* any of those controls can reject it. The downstream adapter cap no longer protects the process that now owns public ingress.

**Required fix:** make the proxy's read bounded by the selected platform's authoritative pre-auth body limit (do not trust `Content-Length` alone; chunked/misreported bodies still need a streaming bound), return `413` before opening the loopback request, and add regressions for all three platforms proving an oversized request never reaches `httpx.AsyncClient.request`.

### P1 — Feishu's source identity collapses to `127.0.0.1`, turning its per-peer rate limit into one attacker-drainable global bucket

Feishu's `_handle_webhook_request()` intentionally derives:

```py
remote_ip = request.remote or "unknown"
rate_key = f"{self._app_id}:{self._webhook_path}:{remote_ip}"
```

and performs that rate-limit check before verification-token/auth processing. The proxy opens a fresh loopback HTTP connection, so every Internet sender arrives at the adapter as `request.remote == 127.0.0.1`.

That means one unauthenticated sender can consume the shared 120/minute bucket and cause legitimate Feishu deliveries from every real peer to receive `429` until the window clears. This is exactly the other side of a reverse-proxy change: payload and signature survive, but transport provenance does not.

**Required fix:** preserve client provenance through a *trusted* proxy contract, not by trusting caller-supplied forwarding headers. For example: strip inbound `Forwarded`/`X-Forwarded-*`, inject a proxy-owned original-peer value, and have the Feishu adapter honor it only when its direct peer is loopback/the trusted local proxy. Add a regression where two distinct public peers remain distinct rate-limit keys, plus a spoofed inbound forwarding header that cannot choose the effective peer identity.

### P1 — `_PORT_BINDING_PLATFORM_PORTS` is display data, not the adapter's listener authority

The new resolver says it finds the adapter's "actual live port" by reusing `_PORT_BINDING_PLATFORM_PORTS` and `_profile_platform_ports()`. Current source says the opposite: that table is explicitly **"Used only for the dashboard's gateway-topology readout — best-effort display data, not a bind source"**, and the existing topology helper explicitly does not resolve env-var port overrides.

That distinction matters here:

- LINE's real listener precedence is `LINE_PORT` -> `extra.port` -> `8646`.
- Feishu's real listener precedence includes `extra.webhook_port` / `FEISHU_WEBHOOK_PORT` -> `8765`.
- `_resolve_webhook_proxy_port()` reads YAML blocks/defaults only.

So a documented `LINE_PORT` or `FEISHU_WEBHOOK_PORT` deployment can have a healthy adapter listening on one port while the new public proxy deterministically dials another and returns `502`. This is not just the already-declared custom-path limitation; it is a mismatch in the port configuration surface the PR currently claims to honor.

**Required fix:** do not reconstruct mutation/routing authority from the topology display projection. Give the listener endpoint one canonical resolver/owned runtime value and have both adapter binding and proxy routing consume it, or at minimum reproduce each adapter's exact effective-port precedence with tests for the documented env overrides. A runtime-owned endpoint is preferable because it removes this drift class rather than adding another copy.

### Architecture / merge-order gate — this grows the exact god-file cluster already under extraction

This PR adds 120 behavior lines directly to `hermes_cli/web_server.py`, the 17,700-line god file tracked by #78628 / #78647. More specifically, open complementary PR #79124 (Axl Ibiza) extracts the **gateway-topology block containing `_PORT_BINDING_PLATFORM_PORTS`, `_PLATFORM_DEAD_STATES`, and the collector/resolver surface this PR is extending** into `hermes_cli/web_gateway_topology.py`. The file lists overlap directly (`hermes_cli/web_server.py`), and this change makes a display-only topology owner serve as security-sensitive ingress routing authority.

I would not solve that by moving this proxy wholesale into `web_gateway_topology.py`; that would preserve the ownership mistake. The clean composition is a focused webhook-ingress/proxy owner (sub-2k), a thin middleware registration in `web_server.py`, and an explicit endpoint contract shared with the platform listener owner. #79124 remains complementary extraction work, not a duplicate or something this PR supersedes; its contributor credit/lineage should stay intact.

### Exact-object release evidence

The PR has one commit, `a837c3cad8f1107b5dd7cd26f863357ed486385d`. Current `main` is `b1ff8722a53ee223485ac9804945acf07ef5c601`, but this branch's merge base is `3f315e46fede84ed4e6c8cfdbd00a13618e68986`: GitHub reports the head **229 commits behind current main**. CI `33252287816`, Nix `33252287517`, and Docker `33252287518` are all `action_required`; the CI run currently has zero jobs. So the reported local 397-test evidence is useful development evidence, but it is not a current-main/exact-head release receipt.

After the boundary fixes above: rebase/recompose on current `main`, rerun the focused dashboard-auth + webhook suites plus the LINE/Feishu/WeCom adapter suites, and require hosted CI/Nix/Docker to actually execute and settle green on that exact commit.

### Non-blocking edge to keep explicit

The same single-public-port topology also affects LINE's public `/line/media/<token>/<filename>` route, which is served by the same adapter and is used when `LINE_PUBLIC_URL` points at a reverse-proxied public host. This PR intentionally proxies only `/line/webhook[/...]`, so it should not claim the whole single-public-port LINE public-route class is closed. That media route can remain follow-up scope; just keep the boundary explicit so the webhook repair is not mistaken for complete adapter-publication coverage.

The core diagnosis here is good, and the live reproduction proves there is something worth landing. The important change is to make the new edge preserve the protections and provenance the old direct-adapter edge already had, rather than treating a reverse proxy as a transparent hop when it is actually a new authority boundary.

PR #93752, opened 2026-08-24
fix(desktop): hide stale workspace on new chat

mrkillbob:
## Summary
- keep the previous cwd as non-authoritative draft state
- hide branch/path and Files tree on the new-chat route
- reveal the workspace only after session.create applies its authoritative cwd

## Verification
- 68 focused desktop UI tests passed after rebasing onto current main
- desktop TypeScript checks passed
- rebuilt macOS app verified with Computer Use: old branch/path/tree absent after New Session

PR #65408, opened 2026-07-16
feat(desktop): re-anchor sessions to projects

wufengwind:
## Summary

- Add a **Move to project…** action to Desktop session row and tab menus.
- Re-anchor the session's authoritative working directory to the selected project instead of introducing a second project-assignment state.
- Route primary and secondary-tab sessions through their live runtime so future tools immediately use the new workspace.
- Update inactive historical sessions through the profile-routed REST API.
- Clear stale branch/repository metadata atomically when changing workspaces.
- Guard asynchronous Git probes by cwd so an older probe cannot overwrite metadata after a rapid second move.

## Motivation

Addresses #54204.

Sessions created outside the intended project currently cannot be reassociated from the Desktop UI. Users must either start over or edit `state.db` manually.

Project membership is already derived from session cwd. This implementation keeps cwd as the single source of truth: moving a session changes both where it appears in the project tree and where future tools execute.

## Safety

- Moving is disabled while the session is running.
- Background-profile sessions cannot be moved using another profile's project list.
- The backend rejects empty or nonexistent directories.
- Open secondary chat tabs are updated through their own runtime.
- Historical sessions are updated without loading their transcript.
- Stale Git metadata is cleared during the move and repopulated using a compare-and-set guard.

## Scope and overlap

This does not introduce an independent project-assignment table or a synthetic **No Project** override. A session belongs to whichever configured project contains its authoritative cwd.

This overlaps with #61335, but deliberately uses cwd-based membership rather than maintaining a separate assignment state. The overlap is called out explicitly so maintainers can choose the desired semantics or salvage the relevant commits without losing contributor context.

## Validation

- `npm run typecheck`
- ESLint on changed Desktop files with zero warnings
- 26 focused Desktop/Vitest tests
- 368 SessionDB tests
- 323 TUI gateway tests
- 392 web server tests
- Ruff checks on changed Python files
- `git diff --check`

PR #49860, opened 2026-06-21
fix: preserve desktop workspace cwd across continuations

kemaldoganay:
## Summary

- Preserve the parent session `cwd` when context compression rotates into a continuation session.
- Separate declared workspace cwd from execution/runtime cwd in TUI gateway session metadata so legacy/no-workspace sessions do not get grouped under the gateway launch directory.
- Hydrate live Desktop sidebar payloads from the stored SessionDB row when a resumed/live session has a persisted cwd but no in-memory explicit marker.
- Sync the Desktop default project dir before resolving cwd for new local sessions and branches.
- Tighten regression coverage for workspace/no-workspace semantics, DB isolation, compression cwd inheritance, Desktop cwd resolution, and an environment-dependent browser hint test.

## Test plan

- `venv/bin/python -m pytest tests/test_tui_gateway_server.py tests/agent/test_compression_logging_session_context.py -q`
  - `289 passed, 8 warnings`
- `npm --prefix apps/desktop run typecheck`
  - passed
- `npm --prefix apps/desktop run test:ui -- src/store/session.test.ts`
  - `21 passed`

## Notes

This fixes a Desktop sidebar drift case where compression continuation sessions could reappear under `No workspace` after app restart even though their parent session belonged to a concrete workspace.

PR #97881, opened 2026-08-29
fix(desktop): preserve authoritative session workspace

vinjaune89:
## Summary

- Separate Desktop execution cwd from authoritative session workspace provenance.
- Preserve `explicit_cwd` through create, resume, branch, compute-host, background, preview, healing, and rebuild carriers.
- Make nested lifecycle ContextVar cleanup stack-safe, including cron workdirs without a session key.
- Emit explicit reset identities and producer capability seals for the Cortex lifecycle adapter.
- Bind classic local CLI turns to the resolved workspace without leaking ContextVar state after the turn.

## Why

Several simultaneous Desktop conversations can share one MCP transport, but they must not inherit workspace ownership from process-global or configured cwd fallback. A fallback path remains useful for tool execution. It is not proof of conversation ownership.

This change keeps those values separate and fails closed when authoritative workspace provenance is absent.

## Verification

- `679` repair-focused tests passed.
- `16` cron workdir/isolation tests passed.
- Ruff passed on all modified Python files.
- Python compilation passed on all modified production files.
- `git diff --check` passed.
- Independent moved-base cold review: **ready**, with no P0–P3 findings.
- Reviewed base: `b1ff8722a53ee223485ac9804945acf07ef5c601`.
- Reviewed full-index diff SHA-256: `d16ca7a29cd66b55295634810036120135c195a4b7c0a777cce97bf99896c092`.

## Baseline limitation

`tests/test_tui_gateway_server.py::test_model_options_preserves_canonical_custom_row_after_agent_init` fails with an unexpected `anthropic` provider row. The same assertion fails unchanged on a clean `b1ff8722a53ee223485ac9804945acf07ef5c601` worktree, so it is not introduced by this patch.

## Scope

This PR changes repository source and tests only. It does not install, restart, deploy, or activate the lifecycle adapter.

### DIFF
--- a/agent/runtime_cwd.py
+++ b/agent/runtime_cwd.py
@@ -46,6 +46,11 @@ def set_session_cwd(cwd: str | None) -> Token:
     return _SESSION_CWD.set((cwd or "").strip())
 
 
+def reset_session_cwd(token: Token) -> None:
+    """Restore the cwd value that preceded a nested session binding."""
+    _SESSION_CWD.reset(token)
+
+
 def clear_session_cwd() -> None:
     _SESSION_CWD.set("")
 
@@ -55,6 +60,15 @@ def _session_cwd_override() -> str:
     if value is _UNSET:
         return ""
     return str(value).strip()
+
+
+def authoritative_session_cwd() -> str:
+    """Return the explicitly bound session cwd, or empty when unproven.
+
+    Lifecycle and plugin attribution must not use ``TERMINAL_CWD`` or the
+    process cwd. One gateway process can serve many concurrent workspaces.
+    """
+    return _session_cwd_override()
 
 
 def resolve_agent_cwd() -> Path:
--- a/agent/turn_context.py
+++ b/agent/turn_context.py
@@ -48,6 +48,7 @@ from agent.model_metadata import (
     estimate_messages_tokens_rough,
     estimate_request_tokens_rough,
 )
+from agent.runtime_cwd import authoritative_session_cwd
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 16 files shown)


alt-glitch:

Related: #65408, #93752, and #49860 address the Desktop workspace/CWD family through different UI, continuation, or ownership mechanisms. This PR's cross-runtime provenance model needs a maintainer comparison before merge.

vinjaune89:
## Relationship to the existing Desktop workspace PRs

Thanks for flagging these. I compared both scope and touched paths:

- **#65408 — explicit “Move to project” action.** This is complementary. It changes the authoritative cwd through a deliberate user action; #97881 governs how that proven value is carried afterward and prevents configured/runtime fallback from being promoted into workspace provenance. The PRs overlap in `tui_gateway/server.py` and `tests/test_tui_gateway_server.py`, so whichever lands second needs an interaction rebase and focused move-session tests.
- **#93752 — hide stale workspace on new chat.** This is complementary and has no file overlap with #97881. It keeps the frontend honest before `session.create`; #97881 enforces the backend provenance boundary when creation and later carriers run.
- **#49860 — continuation and hydration workspace semantics.** This is the closest semantic and file overlap (`tui_gateway/server.py` plus its tests). It establishes part of the same execution-cwd vs authoritative-workspace distinction for continuations/hydration. #97881 extends that invariant across create/resume/branch, compute-host reconstruction, background/preview children, healing, prompt/capability rebuilds, classic CLI turns, reset identities, lifecycle capability seals, and nested `ContextVar` teardown.

My proposed sequencing:

1. Treat #49860 as a partial precursor if it lands first; rebase #97881 and rerun the full exact-hash review.
2. Let #65408 land in either order, but require the second PR to prove explicit moves remain authoritative without reintroducing fallback laundering.
3. Treat #93752 as independently stackable.

Maintainer question: would you prefer #97881 to stack on #49860 as written, or narrow after its continuation/hydration pieces land? I will follow that ownership decision rather than duplicate the same contract in parallel.

vinjaune89:
@teknium1 — requesting maintainer review on the overlap and sequencing question above. The current candidate is independently reviewed against `b1ff8722a53ee223485ac9804945acf07ef5c601`; I will rebase and renew the exact-hash review if an adjacent PR lands first.

PR #97883, opened 2026-08-29
fix(cli): suppress duplicate response panel when TTS displays without token streaming

ziuus:
## Summary

- Fixes a CLI bug where the final response panel duplicates when TTS output is active but token streaming is disabled
- The TTS `display_callback` prints sentences as they're spoken; the final Rich Panel was supposed to be suppressed, but the guard relied on a nonlocal boolean set by the TTS callback in a separate thread — when it failed, the panel rendered on top of TTS output
- Adds an instance variable (`self._tts_displayed_this_turn`) as a reliable, thread-safe indicator; widens the guard to check both the instance var and the local flag
- Normal token streaming path (`streaming_enabled=True`) is unaffected

## Reproduction

This exact conversation showed the bug: running `git push origin master + 4 commands` via Hermes CLI produced the full response twice in terminal output — once from the TTS display callback and once from the Rich Panel rendering the same `response` string.

## Validation

- [ ] Test with TTS enabled, token streaming disabled — confirm no duplicate panel
- [ ] Test with TTS disabled, token streaming enabled — confirm normal streaming behavior unchanged  
- [ ] Test with both enabled — confirm no regression

🤖 Generated with [Claude Code](https://cli.claude.com)

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -16648,6 +16648,10 @@ class HermesCLI(CLIAgentSetupMixin, CLICommandsMixin, CLIBillingMixin):
                 except Exception:
                     pass
 
+            # Track whether TTS displayed content this turn — used to suppress
+            # the final Rich Panel when TTS is the sole display path.
+            self._tts_displayed_this_turn = False
+
             if use_streaming_tts:
                 text_queue = queue.Queue()
                 stop_event = threading.Event()
@@ -16663,6 +16667,7 @@ class HermesCLI(CLIAgentSetupMixin, CLICommandsMixin, CLIBillingMixin):
                     def display_callback(sentence: str):
                         """Called by TTS consumer when a sentence is ready to display + speak."""
                         nonlocal _streaming_box_opened
+                        self._tts_displayed_this_turn = True
                         if not _streaming_box_opened:
                             _streaming_box_opened = True
                             w = self._scrollback_box_width(getattr(self.console, "width", 80))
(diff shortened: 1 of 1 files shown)


Enough1122:

**fix(cli): suppress duplicate response panel when TTS displays without token streaming**

Reasonable fix. The new per-turn `self._tts_displayed_this_turn` flag records that the TTS display callback actually rendered text, and `run_agent` now suppresses the final Rich Panel when TTS was the sole display path (the `tts_already_displayed` condition adds `self._tts_displayed_this_turn or` to the existing `_streaming_box_opened` check). Using a durable per-turn flag in addition to the transient `_streaming_box_opened` state correctly covers the closed-box window.

Non-blocking observations:

1. **Cross-thread flag.** The flag is written from the TTS consumer thread (`display_callback`) and read from the main thread in `run_agent`. For a single bool under the GIL this is fine and the comment calls it thread-safe — but if this ever needs to carry more than a boolean (e.g., a count or a last-display timestamp), switch to a lock or an atomic.

2. **Reset ordering vs. a draining previous turn.** The flag is reset to `False` at the start of `_stage_user_message`. If the *previous* turn's TTS consumer can still be draining (thread not yet joined/stopped) at that moment, a late `display_callback` from the old turn could set the flag `True` after the reset, spuriously suppressing the new turn's response panel. Confirm the TTS thread is joined/stopped before the next `_stage_user_message` runs.

3. **Minor redundancy.** Since the flag is only ever set inside the `use_streaming_tts` branch's `display_callback`, `tts_already_displayed` is effectively `use_streaming_tts and (flag or box_opened) and not is_error_response`. The `or` is still meaningful (box state is transient, flag is durable per turn), so this is a note, not a problem.

Issue #97764, opened 2026-08-29
[Bug] Desktop renderer never resumes stored session after mid-turn ws drop — chat frozen with 'session-scoped RPC rejected ... not in memory' until full app restart


kitanswers:
## Summary

In the Hermes desktop app (macOS, local profiles, Bot Mode with 7 profiles), the renderer's websocket to a profile backend drops mid-turn. The backend correctly persists the turn to the DB, but the renderer never resumes the stored session: every subsequent session-scoped RPC is rejected with `session-scoped RPC rejected: session_id='...' not in memory (detached/reaped runtime; client should resume the stored session)`, and the chat view stays frozen/stale until the user fully quits and reopens the app. Backend-bouncing the profile does not fix it (and can trigger another invalidation the broken recovery path trips over).

Data is never lost — the turn is always in `state.db` — but the UI stops showing it and the next prompt can hang.

## Environment

- Hermes desktop app, macOS 26.x (Mac mini, arm64)
- Local mode, 7 profiles (default + 6 agent profiles, each with its own `hermes serve` backend spawned by the desktop)
- Profile in evidence: `youtube`; main chat session `20260824_112003_b45846` (~2.5k messages, ~175K tokens context)
- Gateway launchd-supervised (`ai.hermes.gateway`)

## Steps to reproduce (intermittent, load-correlated)

1. Open a long-lived desktop chat on a non-default profile (e.g. youtube).
2. Send a prompt. Mid-turn (often during model streaming), the renderer's ws closes with `client_disconnect(code=1005)`.
3. Backend tears down the transport, reaps/detaches the session, interrupts the in-flight turn (`client_gone action=interrupt`).
4. Renderer reconnects within seconds (new ws every ~5s polling loop) but keeps addressing the OLD `ui_session` id.
5. All session-scoped RPCs now return 4001 `session-scoped RPC rejected: session_id='...' not in memory`.

Observed twice within one hour on 2026-08-28/29 (23:32 and 00:25 local), each time on the youtube profile backend. Connect churn that day: 3,742 ws accepts vs 275 the previous day — correlated with heavy machine load (7 Python backends + Electron + long Remotion renders).

## Expected behavior

Per the server-side comment at `tui_gateway/server.py` (around line 2896, referencing #90428): when a session-scoped RPC is rejected because the runtime was detached/reaped, **the client should resume the stored session** on the durable id. The renderer reconnects but never performs that resume, so the view is permanently stale until a full app restart.

## Evidence

[...]

PR #97887, opened 2026-08-29
fix(desktop): recover visible sessions after stale runtime RPC

wz-heng:
## Summary

When Desktop misses a transient WebSocket disconnect and the gateway reaps the runtime session, a later session-scoped RPC can return 4001 even though the durable session still exists.

This routes that signal into the existing route-resume mechanism for the currently visible main session. The original RPC is deliberately not retried: it may be destructive, while the recovery rebinding is asynchronous.

- trigger a durable rebind only for an explicit 4001 (or legacy code-less equivalent) on the selected stored session
- preserve the session owner route for cross-profile/connection recovery
- exclude `session.activate` and `session.resume`, which are the recovery lifecycle itself, so a stale warm-cache activation cannot queue a duplicate resume
- leave tiles and background sessions to their existing recovery lifecycles

Part of #97764.

## Tests

- `npm --workspace apps/desktop run test:ui -- session-rpc-dispatcher.test.ts use-session-actions.test.tsx use-route-resume.test.tsx`
- `npm --workspace apps/desktop exec eslint -- src/app/contrib/session-rpc-dispatcher.ts src/app/contrib/session-rpc-dispatcher.test.ts src/app/session/hooks/use-session-actions.test.tsx`

Snape independently reviewed the final commit and found no remaining blocker, should, or nit.

### DIFF
--- a/apps/desktop/src/app/contrib/session-rpc-dispatcher.test.ts
+++ b/apps/desktop/src/app/contrib/session-rpc-dispatcher.test.ts
@@ -23,10 +23,16 @@ vi.mock('@/store/gateway', async importActual => ({
 }))
 
 const probe = vi.hoisted(() => ({ resolveSessionOwner: vi.fn(async () => undefined as unknown) }))
+const sessionMocks = vi.hoisted(() => ({ requestSessionResume: vi.fn() }))
 
 vi.mock('@/app/session/hooks/use-session-actions/utils', async importActual => ({
   ...(await importActual<Record<string, unknown>>()),
   resolveSessionOwner: probe.resolveSessionOwner
+}))
+
+vi.mock('@/store/session', async importActual => ({
+  ...(await importActual<Record<string, unknown>>()),
+  requestSessionResume: sessionMocks.requestSessionResume
 }))
 
 const { createSessionRpcDispatcher } = await import('./session-rpc-dispatcher')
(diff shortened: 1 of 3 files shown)


Enough1122:

**fix(desktop): recover visible sessions after stale runtime RPC**

Well-scoped and carefully tested. The design is right: on a structured stale-runtime failure (code 4001) for an RPC on the *currently visible* session, request a durable rebind (`requestSessionResume`) but always rethrow the original error (preserving the error contract — it deliberately does not retry a potentially destructive RPC). The exclusions for `session.resume`/`session.activate` (the resume lifecycle itself) and for background rows (via `routingSessionId === selectedStoredSessionIdRef.current`) prevent both double-resumes and foreground route theft. The four new tests cover all the meaningful branches.

Non-blocking observations:

1. **Message-regex fallback can false-positive.** `isStaleRuntimeSessionError` falls back to `/session not found/i` only when there's no numeric code — a reasonable guard for older bridges that lose the code. But any non-coded error whose message happens to mention "session not found" (e.g., a tool output echoed into an error) would be treated as stale-runtime. Impact is limited (only the visible session, only non-resume methods, and the result is a soft `requestSessionResume`), so low risk — flagging for awareness.

2. **Recovery is fire-and-forget.** `requestSessionResume(...)` is not awaited and its failure is not surfaced to the user. If the durable resume also fails, the user sees only the original RPC error with no hint that recovery was attempted and failed. Consider logging the resume outcome (success/failure) for diagnosability.

OutThisLife:
Superseded by #99891.

That salvage keeps requesting a durable resume when the visible session's RPC returns 4001.

You're credited via `Co-authored-by`. Thanks for the dispatcher recovery path.

Issue #89286, opened 2026-08-18
413 payload-too-large recovery loops: text compression runs before image stripping, so base64 vision payloads are never dropped


lurebat:
## Summary

A screenshot-heavy session can wedge in a permanent `413 -> compress -> 413` loop because the payload-too-large recovery path spends every compaction attempt on **text** summarization before it ever tries dropping **image** payloads — and text summarization cannot shrink base64.

## Observed

Real session log (`copilot` / `claude-opus-5`, context window 1,000,000, threshold_tokens 468,000 — so this is NOT a context-window issue):

```
openai.APIStatusError: Request Entity Too Large
API call failed (attempt 1/3) ... summary=HTTP 413: Request Entity Too Large
context compression started: messages=112 tokens=~83,832 model=claude-opus-5
context compression done:    messages=112->51
context compression started: messages=51  tokens=~47,641
context compression done:    messages=51->48
context compression started: messages=48  tokens=~46,261
context compression done:    messages=48->48      <-- zero progress
context compression started: messages=50  tokens=~48,295
```

Each pass cost ~50-70s of wall clock. The session compacted repeatedly at ~84k tokens against a 1M window, because the trigger was the provider's **byte-size** limit, not the token threshold.

## Root cause

In `agent/conversation_loop.py`, the `if is_payload_too_large:` branch is ordered:

1. `compression_attempts += 1`
2. `agent._compress_context(...)`  (slow LLM call)
3. check whether messages/tokens actually shrank
4. **only if step 3 shows no progress** -> `_try_strip_image_parts_from_tool_messages(...)`

Retained vision parts (screenshots, base64 `image_url` blobs) are typically the bulk of an oversized request body. Text summarization leaves them untouched, so steps 1-3 reliably no-op on an image-heavy session and burn one of `max_compression_attempts` per round-trip. Because `_compress_context` legitimately reports "no reduction", the loop can also re-fire from the outer path rather than terminating cleanly.

## Proposed fix

Invert the order inside the `is_payload_too_large` branch: attempt the vision-payload strip **first** (it is instant, local, and targets the actual bytes), and fall through to text compression only when there was no image payload left to drop.

```python
if is_payload_too_large:
    if agent._try_strip_image_parts_from_tool_messages(
        api_messages, remember_model=False,
    ):
        agent._buffer_status(
            "Request payload too large (413) - dropped retained "
            "vision payloads and retrying before compressing..."
        )
        continue

    compression_attempts += 1
    ...
[...]

PR #97889, opened 2026-08-29
fix(agent): evict stale screenshot payloads before send

### LINKED ISSUE #89296: Screenshots are re-uploaded on every turn: evict stale vision payloads before send instead of after a 413
## Summary

Follow-up to #89286. That issue is about recovering from a 413 *correctly*; this one is about not provoking it in the first place.

Base64 image parts from screenshot tools are re-serialized and re-uploaded on **every subsequent API call** for the remainder of the conversation, even though only the most recent one or two carry decision-relevant signal. In a browser or computer-use session this quietly becomes the dominant cost in both bytes and latency, and eventually trips the provider's payload limit — at which point the reactive path strips the images anyway, *after* the oversized body has already been serialized and sent.

The cheapest byte is the one never sent.

## Measurement

Six screenshot-bearing tool results (a very ordinary browser session):

```
payload 1,202,057 -> 401,917 bytes  (66.6% smaller)
```

That reduction is paid on *every turn* thereafter, not once.

## Proposal

Evict stale vision payloads on the outbound path, right after `_sanitize_api_messages()`, keeping the N most recent image-bearing tool messages intact and downgrading older ones to their text parts plus an explicit marker:

```
[older screenshot evicted from context to save payload size]
```

The text signal survives, so the model still knows a screenshot was taken and what it showed — it just can't re-examine an old one. In practice the agent has already acted on it.

Key design points:

- **Only `role: "tool"` messages are eligible.** User-attached images are deliberate context the user may refer back to at any point; silently dropping them would change the meaning of the conversation rather than trimming redundant agent scratch data.
- **Configurable** via `compression.vision_keep_last_images` (default `2`). `0` evicts every image; the value is clamped to `[0, 20]`. Note the floor is 0, not 1, unlike `compression.max_attempts` — "keep no images" is a meaningful setting.
- **Fail-open.** The call is wrapped so an eviction bug can never break the request path.
- Complements rather than replaces the 413 handler: the reactive strip stays as the safety net for the case where even the retained images are too large.

## Implementation

New module `agent/image_eviction.py` exposing `evict_stale_image_parts(api_messages, *, keep_last=2) -> int`, invoked from the outbound request path in `agent/conversation_loop.py`.

Tested with 11 unit tests covering: recency ordering, text-signal preservation, `keep_last=0`, negative clamping, user-image immunity, image-free messages, and non-list input as a safe no-op.

## Environment

- `main` @ 9664e386f6
- Observed originally on provider `copilot`, model `claude-opus-5`, but the behavior is provider-independent.

Christopher-Schulze:
## What does this PR do?

Stops stale `vision_analyze` / screenshot tool results from being re-serialized on every later API call.

Compression already retires older image-bearing tool payloads, but only when prune/compress fires. The Anthropic adapter has a keep-window too, but it only sees nested `tool_result` image blocks. OpenAI-style `image_url` tool results therefore rode every subsequent turn until a 413 forced the reactive strip (#89286).

The existing keep-newest helper now runs on the per-call `api_messages` clone immediately after sanitization. Persisted history is untouched. The keep window is the same `_MAX_KEEP_TOOL_IMAGES` (3) used by compression.

## Related Issue

Fixes #89296

## Type of Change

## Changes Made

- `agent/context_compressor.py`: public `evict_stale_outbound_tool_images()` wraps the existing keep-newest retirement for the send path.
- `agent/conversation_loop.py` and `agent/chat_completion_helpers.py`: call it on the cloned API list after `_sanitize_api_messages`.
- `tests/agent/test_outbound_stale_vision.py`: sanitizer-alone still keeps every screenshot (the previous chokepoint); eviction keeps the newest window; clones do not rewrite persisted history; user uploads are not touched.

## How to Test

```bash
scripts/run_tests.sh tests/agent/test_outbound_stale_vision.py tests/agent/test_compressor_stale_tool_images.py
```

Final local head `fd9e2ca49d22da4d0cc76ecb6f49801c58bd6e36`: 4/4 new tests passed; `scripts/check.sh --project hermes-agent --worktree worktrees/hermes-agent/89296` passes all blocking gates.

## Checklist

- [x] Conventional commits, scoped, tests pass, macOS

### DIFF
--- a/agent/chat_completion_helpers.py
+++ b/agent/chat_completion_helpers.py
@@ -3270,6 +3270,10 @@ def handle_max_iterations(agent, messages: list, api_call_count: int) -> str:
         # tool_call was summarized away; Responses API rejects that as
         # "No tool call found for function call output".
         api_messages = agent._sanitize_api_messages(api_messages)
+        # Same send-path vision eviction as the main loop (#89296).
+        from agent.context_compressor import evict_stale_outbound_tool_images
+
+        evict_stale_outbound_tool_images(api_messages)
 
         # Same safety net as the main loop: drop thinking-only assistant
         # turns so Anthropic-family providers don't 400 the summary call.
--- a/agent/context_compressor.py
+++ b/agent/context_compressor.py
@@ -1772,6 +1772,23 @@ def _retire_stale_tool_result_images(
     return pruned
 
 
+def evict_stale_outbound_tool_images(
+    api_messages: List[Dict[str, Any]],
+    keep_newest: int = _MAX_KEEP_TOOL_IMAGES,
+) -> int:
+    """Drop stale screenshot/vision payloads from the per-call API copy.
+
+    Compression's keep-newest pass only runs when prune/compress fires, and
+    the Anthropic adapter's screenshot eviction only sees nested
+    ``tool_result`` blocks. OpenAI-style ``image_url`` tool results
+    otherwise ride every subsequent request until a 413 forces the reactive
+    strip (#89286). Call this on the cloned ``api_messages`` list after
+    sanitization so older frames never leave the box (#89296). Do not pass
+    persisted history — the rewrite is send-path only.
+    """
+    return _retire_stale_tool_result_images(api_messages, keep_newest=keep_newest)
+
+
 def _truncate_tool_call_args_json(args: str, head_chars: int = 200) -> str:
     """Shrink long string values inside a tool-call arguments JSON blob while
     preserving JSON validity.
--- a/agent/conversation_loop.py
+++ b/agent/conversation_loop.py
@@ -2743,6 +2743,13 @@ def run_conversation(
         # gated on context_compressor — so orphans from session loading or
         # manual message manipulation are always caught.
         api_messages = agent._sanitize_api_messages(api_messages)
+        # Send-path vision eviction (#89296): compression only strips stale
+        # screenshots when prune fires, and the Anthropic adapter's keep-window
+        # never sees OpenAI-style tool-result image_url parts. The per-call
+        # clone is rewritten in place; persisted history is untouched.
+        from agent.context_compressor import evict_stale_outbound_tool_images
+
+        evict_stale_outbound_tool_images(api_messages)
 
         # One-time repeated-heal escalation notice (#96870): if the sanitizer
         # above just crossed the per-session heal threshold, deliver the
--- a/tests/agent/test_outbound_stale_vision.py
+++ b/tests/agent/test_outbound_stale_vision.py
@@ -1,0 +1,119 @@
+"""Send-path eviction of stale vision_analyze / screenshot tool payloads.
+
+Issue #89296: compression only retires older image-bearing tool results when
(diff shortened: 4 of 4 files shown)


Enough1122:

**fix(agent): evict stale screenshot payloads before send**

Good fix for #89296. Wrapping `_retire_stale_tool_result_images` as `evict_stale_outbound_tool_images` and calling it at the per-call send chokepoints (`run_conversation` and `_managed_summary_call`, both right after sanitize) stops OpenAI-style `image_url` screenshot payloads from riding every subsequent request until a 413 forces a reactive strip. The comment correctly notes the existing paths miss this (compression's keep-newest only runs on prune/compress; the Anthropic adapter's eviction only sees nested `tool_result` blocks).

The tests are solid: `test_eviction_keeps_only_newest_window` (keep window + text marker), `test_does_not_rewrite_persisted_history`, and `test_user_uploads_are_not_evicted` (user-uploaded images are never touched — an important safety property).

Non-blocking observations:

1. **History-leak safety verified.** I confirmed the send path builds `api_messages` via `_clone_message_for_send` (a structural deep clone that decouples nested containers), so the in-place eviction cannot rewrite persisted history — consistent with the function's "send-path only" contract and the `test_does_not_rewrite_persisted_history` pin. Good.

2. **Thin wrapper.** `evict_stale_outbound_tool_images` is a pure passthrough to `_retire_stale_tool_result_images` with a default `keep_newest`. Fine as a named, documented chokepoint; just noting the indirection is intentional naming, not new logic.

3. **Visible text marker.** The eviction replaces stripped images with a text marker containing "screenshot removed" (per the test). That text is what the model sees in the transcript — worth confirming the wording is intentional and clear to the model (it is a reasonable "this image was dropped" signal).

kshitijk4poor:
**Triage note (perf/P2 sweep)** — checked specifically through the prompt-cache lens (it touches `context_compressor` + `conversation_loop` outside the compression path).

Probe on `origin/main` 1cb3ab6173 with the PR applied: history of 6 screenshot tool results, `api_messages` built via `_clone_message_for_send` (containers cloned, leaves shared), then `evict_stale_outbound_tool_images(api_messages)`:
- persisted history after the call: **byte-identical** (`deepcopy` compare) — `_strip_images_from_tool_msg` builds a new dict, nothing is mutated in place on stored messages;
- images sent per request: 6 → **3** (keep-newest);
- prefix stability: when a 7th screenshot arrives, the request diverges from the previous one only at the first evicted message (index 9 of 14 in the probe) — the same shape as compression's keep-newest, so the cache break is bounded to the eviction boundary and intentional, not a whole-prefix invalidation.

4/4 tests; needed one context-only conflict in `agent/conversation_loop.py` against current main. No cache-rule violation found.

kshitijk4poor:
Merged via #101709 (rebase — your commit landed on `main` as cf86c47624 with authorship preserved; I only resolved a `conversation_loop.py` context conflict). Thank you @Christopher-Schulze — stale screenshots are evicted from the outbound copy only, history byte-identical, prefix stable up to the eviction frontier.

Issue #89298, opened 2026-08-18
Desktop-spawned `hermes serve` backends leak unbounded (28 live instances); leaked backends' MCP OAuth storms browser login tabs


ninearif:
# Desktop-spawned `hermes serve` backends leak unbounded; leaked backends' MCP OAuth storms browser login tabs

## Summary

The desktop app's headless backends (`hermes serve`) can accumulate without bound — I found **28 live `serve` processes for one profile**, all orphaned (PPID 1). Each one eagerly connects configured MCP servers at startup, so an OAuth-based stdio server (`npx -y mcp-remote https://mcp.atlassian.com/v1/mcp`) spawned 6+ concurrent proxies that all raced on mcp-remote's fixed OAuth callback port (3736). No flow could complete, every retry opened a fresh browser tab, and the user experience was 10+ Atlassian authorization tabs popping up continuously.

## Environment

- macOS 15 (Darwin 24.6.0), Apple Silicon
- hermes-agent 0.20.1, checkout `b2369172a` (2026-08-16)
- Hermes Desktop (managed install at `~/.hermes/hermes-agent`), profile in use: a named profile via `--profile`
- MCP server configured in profile `config.yaml`:

```yaml
mcp_servers:
  atlassian:
    command: npx
    args: [-y, mcp-remote, "https://mcp.atlassian.com/v1/mcp"]
```

## What I observed

- Browser repeatedly opened `mcp.atlassian.com/v1/authorize?...` tabs ("MCP CLI Proxy is requesting access"), 10+ within minutes.
- `pgrep -fc hermes_cli.main` → **28**, all `python -m hermes_cli.main --profile <name> serve --host …`, all parented to launchd (PPID 1) — i.e., their desktop parent was long gone.
- Six of them had live `tools/mcp_stdio_watchdog.py → npx -y mcp-remote https://mcp.atlassian.com/v1/mcp` children; one held `127.0.0.1:3736` (mcp-remote's OAuth callback) and the others were connected to it, so the auth flows deadlocked/raced and retried forever, opening a tab each time.
- Killing an mcp-remote child had no lasting effect — the running backends respawn it (as designed).
- `~/Library/Application Support/Hermes/backend-ownership.json` **did not exist** while all 28 backends were alive.

## Root cause analysis (from reading the code)

[...]

PR #97892, opened 2026-08-29
fix(desktop): clear stale macOS backend ownership records

foras910521-lab:
## Summary

On macOS, the process identity probe uses ps -p. When the PID no longer exists, ps exits with status 1 and Node reports error.code as numeric 1. The existing code only treated ENOENT and ESRCH as definitive absence, so a dead backend or parent stayed indeterminate and its ownership record survived every reap.

This change:

- recognizes ps exit status 1 as a definitive missing-process result on Darwin only
- applies the same rule to backend and parent identity probes
- keeps non-Darwin exit status 1 conservative
- covers numeric and string status forms with a pure behavior test

This is a narrow bookkeeping fix related to #89298. It does not claim to solve ownership-file loss or scan for unrecorded live orphan processes.

## Validation

- Electron Vitest project: 109 tests passed
- npm audit: 0 vulnerabilities
- signed macOS arm64 Desktop package and install succeeded
- real node-pty smoke passed in source, packaged, and installed-app contexts

### DIFF
--- a/apps/desktop/electron/backend-ownership.test.ts
+++ b/apps/desktop/electron/backend-ownership.test.ts
@@ -7,7 +7,8 @@ import {
   type BackendIdentity,
   createBackendOwnership,
   createBackendShutdownCoordinator,
-  parseBackendOwnership
+  parseBackendOwnership,
+  processProbeConfirmsMissing
 } from './backend-ownership'
 
 function memoryStore(initial = '') {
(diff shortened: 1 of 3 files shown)


Enough1122:

Reviewed PR #97892 (`fix(desktop): clear stale macOS backend ownership records`).

The change centralizes the process-probe "missing process" decision in `processProbeConfirmsMissing` and adds the Darwin-only interpretation of a numeric exit code `1` from `ps -p <pid>` (Node surfaces it as a numeric `error.code` rather than `ESRCH`). The platform gating is the right call, and the test table covering darwin/linux × numeric/string codes is good coverage.

Non-blocking observations:

- **`code === '1'` string form is untested but harmless.** The test only exercises `code: 1` (number) on darwin; the string `'1'` branch is covered implicitly by the `code: '1'` assertion on line 21 — actually that line covers it. Good.
- **Semantics of exit code 1 on macOS.** Treating `ps` exit 1 as definitive "missing" is only safe because the probe command and args are fixed (`ps -p <pid>`). If the probe were ever changed (e.g. adding flags that could exit 1 for other reasons), this would silently mislabel a process as dead. Worth a brief comment near the probe call site (or a guard) tying the numeric `1` interpretation to the current `ps -p` invocation — the docstring already implies this, so this is purely defensive.

No blocking issues. The fix is focused and well-tested.

OutThisLife:
Superseded by #99866.

Same fix class: macOS `ps -p` exits 1 on a missing PID, so stale ownership records survived every reap. The salvage throws ESRCH from a native liveness check before `ps` runs, so the Darwin-only exit-1 mapper is no longer needed.

You're credited via `Co-authored-by`. Thanks for catching the Darwin side of this.

PR #97893, opened 2026-08-29
fix(desktop): position find bar below titlebar to not overlap window controls

### LINKED ISSUE #97890: [Desktop/Windows] Ctrl+F find bar renders over the titlebar controls (overlaps top-right buttons)
Summary
On the Windows desktop app (packaged Electron build, Hermes v0.20.0), pressing Ctrl+F opens the built-in find-in-page bar at the very top of the window (~8px from the top edge), where it overlaps the app's own top-right titlebar controls and is awkward to use. The same issue is still present on current main — the recent find-bar rewrite did not address it.

Environment
OS: Windows 10
Hermes Agent v0.20.0 (2026.8.3), packaged desktop app (Electron)
Confirmed by code inspection on latest main
Root cause
The find bar is positioned with:

apps/desktop/src/components/find-bar.tsx

'pointer-events-auto fixed right-4 top-[calc(var(--titlebar-height,0px)+0.5rem)] z-50'

But the whole app is wrapped by `ContribController`
(`apps/desktop/src/app/contrib/controller.tsx`), which explicitly sets
`--titlebar-height: 0px`:

```tsx
style={{ '--titlebar-height': '0px' } as CSSProperties}


while the real titlebar is `h-[34px]` right below it. The find bar mounts
inside this subtree (`ContribWiring` → `<FindBar />`), so the CSS variable
resolves to `0px`, and the bar renders at `top: 0.5rem` — i.e. inside the
34px titlebar strip, exactly over the top-right controls.

Note: `main` bumped the fallback to `var(--titlebar-height,34px)`, but that
does not help here, because the variable is explicitly set to `0px`
(explicit values win over fallbacks).

## Expected behavior
The find bar should render below the titlebar so it never covers the
titlebar controls.

## Suggested fix direction
- Stop zeroing `--titlebar-height` in the contrib layout, or make it match
  the real titlebar height (a shared `TITLEBAR_HEIGHT` constant instead of a
  hardcoded `34px` here and `0px` there); or
- Anchor the find bar to a real layout element rather than a CSS var that is
  zeroed inside its own mounting subtree.

## Workaround (temporary)
A desktop plugin can reposition the bar via injected CSS
(`div[role="search"] { top: 3rem !important; }` — `role="search"` is unique
to the find bar). Confirms the element is positionable via CSS; stopgap only.

---

shellybotmoyer:
## Summary

Fixes the Ctrl+F find bar overlapping the titlebar window controls on Windows/Linux (issue #97890).

### Root cause

The find bar (`FindBar`) mounts inside the `ContribController` subtree, which explicitly sets `--titlebar-height: 0px` on the app shell div (`controller.tsx` line 841). The find bar's className uses `top-[calc(var(--titlebar-height,34px)+0.5rem)]` — the `34px` fallback was meant to handle this case, but an explicit `0px` value on an ancestor wins over the fallback in the `var()` function. So the bar renders at `top: 0.5rem`, inside the 34px titlebar strip, overlapping the native min/max/close controls.

### Fix

Set `--titlebar-height` to `TITLEBAR_HEIGHT` (34px) on the find bar's root `<div>` via the inline `style` prop. This scopes the override to just the find bar element — no other elements that depend on the `0px` value are affected.

This is the simplest and most targeted fix. Alternatives considered:
- Changing `--titlebar-height` from `0px` to `34px` in `controller.tsx` would affect every element inside the app shell that uses the var for layout offsets (sidebar padding, thread list padding, right sidebar `pt`, etc.) — too risky.
- Hardcoding `34px` in the find bar className instead of using the CSS var would lose the single source of truth for the titlebar height.

Closes #97890

### DIFF
--- a/apps/desktop/src/components/find-bar.tsx
+++ b/apps/desktop/src/components/find-bar.tsx
@@ -1,8 +1,10 @@
 import { useStore } from '@nanostores/react'
 import { useEffect, useRef, useState } from 'react'
+import type { CSSProperties } from 'react'
 import { useLocation } from 'react-router'
 
 import { appViewForPath, isOverlayView } from '@/app/routes'
+import { TITLEBAR_HEIGHT } from '@/app/shell/titlebar'
 import { Tip } from '@/components/ui/tooltip'
 import { useI18n } from '@/i18n'
 import { findBarKeyAction, formatMatchLabel } from '@/lib/find-in-page'
@@ -278,7 +280,15 @@ export function FindBar() {
 
   const matchLabel = formatMatchLabel(query, matchOrdinal, matchCount)
 
-  const barStyle = filesPaneRight != null ? { right: `calc(${filesPaneRight}px + 0.75rem)` } : undefined
+  const barStyle: CSSProperties = {
+    // The find bar mounts inside the ContribController subtree, which sets
+    // --titlebar-height: 0px on the app shell div. Without overriding it here,
+    // the bar renders at top: 0.5rem — inside the 34px titlebar strip,
+    // overlapping the window controls (issue #97890). Set the var to the real
+    // titlebar height so the top offset clears the titlebar band.
+    '--titlebar-height': `${TITLEBAR_HEIGHT}px`,
+    ...(filesPaneRight != null ? { right: `calc(${filesPaneRight}px + 0.75rem)` } : {})
+  }
 
   return (
     <div


Enough1122:

Reviewed PR #97893 (`fix(desktop): position find bar below titlebar to not overlap window controls`).

The change restores the real `--titlebar-height` on the find bar's own style so it clears the titlebar band, overriding the `0px` value injected by the ContribController subtree. Scoping the override to the bar element itself is clean and avoids leaking the fix.

Non-blocking observations:

- **Verify the CSS consumer reads the variable.** The diff sets `--titlebar-height` on the bar, but the consuming rule (the one that computes `top: calc(var(--titlebar-height) + 0.5rem)` or similar) is outside this diff. If any rule hardcodes `top: 0.5rem` or reads a different variable, the override silently does nothing. Worth a quick check that the find-bar styles actually consume this var (this looks like it's the case per the linked issue, but it's not visible in this change).
- **Minor:** `...(filesPaneRight != null ? { right: ... } : {})` — the empty-object branch is unnecessary since spreading `{}` is a no-op; `filesPaneRight != null ? { right: ... } : undefined` would read slightly cleaner, but this is purely stylistic.

No blocking issues.

Issue #87768, opened 2026-08-16
[Feature]: Desktop plugin SDK — per-message render slots (message footer / action bar)


cjjdaq:
### Problem

The desktop plugin SDK exposes application-level surfaces only: `statusBar.left/right`, `panes`, `PALETTE_AREA`, `KEYBINDS_AREA`, `ROUTES_AREA`, `SIDEBAR_NAV_AREA`, and titlebar areas. There is **no per-message slot** inside the thread: a plugin cannot render a component into an assistant message's footer or action bar. The message subtree (`AssistantMessage`, `AssistantActionBar` in `apps/desktop/src/components/assistant-ui/thread/`) is hardcoded core UI with no extension point.

### Use case

A plugin wants to show per-turn usage stats (duration, input/output tokens, cache hit %, completion time) under each assistant message — a common pattern in ChatGPT/Claude-style UIs. This is currently impossible through a supported path. The workarounds we actually tried:

1. **Inject DOM nodes into the message subtree** — React reconciliation removes them on any re-render (message settle, virtual list scroll, state changes). Verified empirically across 7 plugin versions: injected nodes survive a moment, then vanish (`inject ok=true` but the node is gone on the next reconcile; even core-rendered siblings get detached when a message root is remounted).
2. **Fixed-position floating overlay anchored to message coordinates** via `getBoundingClientRect` — works but is fragile (transform containers, nested scroll containers, zoom) and duplicates layout logic the core already owns.

Neither is a supported extension path. `ctx.register` has no message-scoped area, and `Contribute` is page-scoped, not message-scoped.

### Proposed API (suggested shape)

Something like:

```js
ctx.register({
  id: 'msg-stats',
  area: 'messageFooter',   // NEW: renders inside each assistant message footer
  order: 10,
  render: (msgCtx) => /* msgCtx: { messageId, role, isComplete, getText() } */
})
```

The host would render the component inside the existing footer with the same lifecycle as `AssistantActionBar`, so React owns it — no DOM hacks, survives re-renders and remounts by construction.

### Related

- #49363 (desktop dashboard plugin loading — app-level)
- #15658 (page-scoped plugin slots — a merged pattern that could be extended to message scope)
- #72854 (tool-card output visibility — different surface, same theme of "plugin content inside the thread")

### Environment

- Hermes Desktop, latest main
- Plugin SDK per `skills/hermes-desktop-plugins/SKILL.md`

### Labels

comp/desktop · type/feature · plugin-system

Issue #9527, opened 2026-04-14
[Feature]: Add display.show_output_speed to show tokens-per-second after each response


fangweisi:
## Problem or Use Case

I am using Hermes Agent and pay close attention to model response speed. I would like to automatically display the average token output speed after each response (e.g. 45.2 tok/s), so I can more intuitively understand model performance. Currently display.show_cost can show cost, but there is no similar option to display speed.

## Proposed Solution

Add a new configuration option in config.yaml:

```yaml
display:
  show_output_speed: true   # Show average token output speed after response
```

Implementation approach:
1. Add output_speed to the result dict in run_agent.py (near the final return, refer to api_duration and completion_tokens)
2. In cli.py response display logic, read this value and display it

This change is relatively small, belonging to the display.* configuration series, consistent with the existing show_cost style.

## Alternatives Considered

- Previously considered using third-party tools for monitoring, but not intuitive enough
- Also considered using /status command, but wanted automation
- Ultimately, adding output_speed field to result is the cleanest approach - downstream (CLI/Gateway) can consume it as needed

## Feature Type
Configuration option

## Scope
Small (single file, < 50 lines)

## Contribution
I would like to implement this myself and submit a PR

GottZ:
> *This was generated by AI during triage.*

#### Summary

Two PRs are associated with this issue complex: #26932 adds a gateway-footer TPS field but does not implement the requested `display.show_output_speed` CLI option and measures cumulative session output over tool-inclusive turn time, while #44878 implements the option and derives output speed from matching per-provider-call tokens and API duration.

#### Related pull requests

- #26932 `unrelated` — (+189/-0) — unrelated to #9527 and partial for the related footer request: it adds an opt-in gateway `tps` field, but omits `display.show_output_speed` and pairs cumulative session tokens with whole-turn elapsed time. Despite the contributor `keep_open` review on #26932, #44878 directly implements that review's requested matching per-request usage and `api_duration`, so #26932 is superseded.
- #44878 `best fix` — (+158/-6) — best existing fix: it records output tokens and duration from the same successful provider call, adds `display.show_output_speed`, and exposes the resulting rate through the CLI, desktop status bar, and runtime footer. Its broad diff also includes `/tps` and slash-worker/RPC changes beyond the core issue.

#### Duplicates

[...]

PR #97894, opened 2026-08-29
feat(desktop): show token rate on completed assistant replies

Aavirash:
## What
Adds an opt-in (on by default) footer readout on completed assistant messages that shows the **exact provider token count and tokens/sec** for the turn, including multi-step agentic turns. A new toggle in **Settings → Appearance → "Show token rate"** controls it.

## Why
Local-model users (LM Studio / MTPLX) and anyone watching generation speed want a precise, always-visible tok/s without reaching for a separate menu-bar tool. The readout renders only once the turn is complete, so it sits at the bottom of the reply and reflects the whole turn — not a mid-stream snapshot.

## How it works
- The gateway sends session-cumulative `usage`; `message-stream.ts` snapshots it at `message.start` and forwards the **per-turn delta** to `completeAssistantMessage`, which stamps `ChatMessage.usage`.
- `chat-runtime.ts` persists `usage` to message metadata so it survives rehydration / session switches.
- When the provider reports `usage.output`, the count is exact. When it doesn't, the chip honestly shows `≈` (a chars/4 estimate of visible text) — never a fake exact number.
- The chip is gated on turn completion and on the new `$showTurnPerf` store (default `true`).

## How to test
1. Settings → Appearance → toggle **Show token rate** off → replies show no chip; on → chip returns (live, no restart).
2. Send a normal reply → `⚡ N tokens · M tok/s` at the bottom, exact (no `≈`) for models that report usage.
3. Run an agentic turn (write → tool → write → tool → write) → chip appears only at the very end after all text/tools.
4. Switch profiles/sessions and return → chip persists (usage is saved to message metadata).
5. Switch the app language (en/ja/zh/zh-hant) → toggle label + tooltip localize.

## Platforms tested
- macOS (arm64) — desktop Electron app.

## Notes
- Icon: Lucide `Zap` via `@/lib/icons` (DESIGN.md icon rule).
- All user-facing strings go through `useI18n()` in en/ja/zh/zh-hant.
- `check-windows-footguns.py`: clean. `tsc --noEmit`: clean. `vitest` (use-message-stream suite): 141 passing, including 3 new cases for per-turn usage forward/delta/omit.

## Contributor
Mapped `aavirash@gmail.com -> aavirash` (satisfies the Contributor Attribution Check).


0xibis:
+1 for this landing by default. I wanted exactly this on desktop: a CLI-style `⏱ 12.4s · ↓ 1.2k tok · 48 tok/s` line under each finished reply. I'm running it as a local desktop plugin for now, but that has to inject into the DOM because there's no per-message render slot (#87768 was closed without one landing on `main`). A native version like this PR would be much more robust. Related: #9527, #118405, #115659.

PR #97895, opened 2026-08-29
feat(discord): add DISCORD_DISABLE_DMS for public community bots

frahlg:
## Summary
Community/support Discord bots often need **public channels only** (no DMs). The Slack adapter already has `SLACK_DISABLE_DMS`; Discord had no first-class equivalent, so local adapter patches were overwritten by `hermes update`.

This adds `DISCORD_DISABLE_DMS` / `discord.disable_dms`:
- turn off the `dm_messages` intent
- drop DMs at admission and in `_handle_message`
- reject slash commands invoked in DMs with a short ephemeral
- leave guild `@mentions` unchanged

## Test plan
- [x] Admission helper: extra/env true → DMs dropped, guild mention still admitted
- [ ] `scripts/run_tests.sh tests/gateway/test_discord_disable_dms.py`
- [ ] With `DISCORD_DISABLE_DMS=true`, DM the bot → no reply; `@mention` in a guild channel still works


Enough1122:

Reviewed PR #97895 (`feat(discord): add DISCORD_DISABLE_DMS for public community bots`).

The feature is well-scoped: gate the `dm_messages` intent at connect, drop DMs in the message admission and `_handle_message` paths, reject slash commands in DMs with a user-facing ephemeral reply, and wire up env/config + docs + tests. Defense in depth across intent and admission is reasonable (the intent alone would not cover interaction events, which arrive regardless of message intents, so the explicit slash-command check is necessary).

Non-blocking observations:

- **Two divergent "is a DM" definitions.** `_discord_message_is_dm` treats a message as DM when the channel is a `DMChannel` *or* `guild is None` (which also catches `GroupChannel`), but `_evaluate_slash_authorization` only checks `isinstance(chan_obj, discord.DMChannel)`. A group DM slash command would therefore not be rejected when DMs are disabled, while group DM *messages* would be. Consider reusing a single helper (e.g. a channel-based `_is_dm_channel` that also handles `GroupChannel`) for both call sites so the two paths stay consistent.
- **`raw.strip().lower() in {"true","1","yes","on"}`** — the string truthiness list differs slightly from other Discord gates in the codebase (some accept `"y"`). Harmless, but worth confirming the accepted values match the docs table wording.
- Minor: `getattr(discord, "DMChannel", None)` is defensive to a fault — `discord.DMChannel` always exists in the installed library; the fallback branch is dead code in practice. Fine either way.

No blocking issues; tests cover the DM-drop and guild-mention-preserved paths.

PR #97900, opened 2026-08-29
fix(desktop): preserve batch clarify choices after skip

adzer000:
## Summary
- preserve authoritative batch `choices_offered` after a blank/skipped response
- render late choices only for unanswered rows
- keep answered rows free of late-choice actions

## Verification
- focused UI contract: 38/38 passed on exact delivery commit
- ESLint: 0 errors (20 existing warnings)
- `git diff --check`: passed
- delivery base: latest upstream `5831d8365aea36fea48f9c3e839827d19ccb9a6c`

Brain was unavailable; no Brain-backed evidence is claimed.

### DIFF
--- a/apps/desktop/src/components/assistant-ui/clarify-tool.test.tsx
+++ b/apps/desktop/src/components/assistant-ui/clarify-tool.test.tsx
@@ -700,6 +700,64 @@ describe('ClarifyTool batch card', () => {
     })
   })
 
+  it('keeps choices actionable after a one-question batch skip', async () => {
+    const inserts: string[] = []
+
+    const stop = onComposerInsertRequest(detail => {
+      inserts.push(detail.text)
+    })
+
+    try {
+      renderClarify(
+        <ClarifyTool
+          {...settledClarifyProps(
+            { questions: [{ choices: ['red', 'blue'], question: 'Color?' }] },
+            {
+              responses: [
+                { choices_offered: ['red', 'blue'], question: 'Color?', user_response: '' }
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed PR #97900 (`fix(desktop): preserve batch clarify choices after skip`).

The fix surfaces the original `choices_offered` on skipped batch rows so the user can still pick a late answer after the batch was skipped, and falls back to the args-declared questions/choices when the response row omits them. The `useMemo` around `readClarifyArgs` and the `row.choices === undefined ? fallback : row.choices` precedence are both sensible.

Non-blocking observations:

- **Index alignment between `responses` and `fallbackQuestions`.** The fallback lookup is positional (`fallbackQuestions?.[index]`). If the settled batch response ever reorders, drops, or dedupes rows relative to the args order, the fallback question/choices would silently attach to the wrong row. The args order and response order appear to match today (and the new tests cover the aligned case), so this is defensive: consider keying by `id` when present instead of position if a mismatch is possible.
- **Empty-choices edge case.** When `choices_offered` is present but empty, `readClarifyBatchResult` stores `null` and the row renders no late choices even though the fallback args may still have choices. That's a deliberate "authoritative empty" signal, but it's worth being aware that a provider returning `choices_offered: []` will suppress the fallback choices entirely. If that's unintended, treating `[]` as "unknown" (falling back) rather than "no choices" would be safer.

No blocking issues; the new tests (skipped row keeps actionable choices; answered row renders none) directly cover the regression.

PR #97901, opened 2026-08-29
fix(plugins): report effective bundled status

Ratzzz33:
## Summary

- report bundled backend and platform plugins as enabled by default, matching PluginManager runtime behavior
- keep explicit disable precedence, opt-in standalone/entry-point semantics, and remove filesystem model-providers from the generic toggle surface their loader does not honor
- make CLI list/info, interactive toggle, plugins.manage, and the web plugins hub use the same effective status
- preserve config on no-op/ESC and persist only actual checkbox deltas
- document the distinct bundled, dashboard, and model-provider activation paths

## Tests

- scripts/run_tests.sh on the four relevant files: 41 passed
- uv run ruff check on all touched Python files
- git diff HEAD^ --check

An independent review initially caught destructive no-op toggle persistence and web-hub status drift; both were fixed and the final review plus evidence verifier passed.

### DIFF
--- a/hermes_cli/plugins_cmd.py
+++ b/hermes_cli/plugins_cmd.py
@@ -1805,10 +1805,11 @@ def _is_portable_plugin_dir(dir_path) -> bool:
         return False
 
 
-# Manifest kinds that are active-by-default when bundled: backends auto-load,
-# platforms register lazily but are available out of the box, model providers
-# run through providers/ discovery (see PluginManager.discover_and_load).
-_BUNDLED_DEFAULT_ON_KINDS = frozenset({"backend", "platform", "model-provider"})
+# Manifest kinds whose bundled activation is owned by PluginManager: backends
+# auto-load and platforms register lazily.  Filesystem model providers use the
+# separate providers/ discovery path, which does not honor plugins.disabled,
+# so they must not be represented as toggleable general plugins here.
+_BUNDLED_DEFAULT_ON_KINDS = frozenset({"backend", "platform"})
 
 
 def _bundled_default_on(dir_path) -> bool:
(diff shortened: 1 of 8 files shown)


Enough1122:

Reviewed PR #97901 (`fix(plugins): report effective bundled status`).

The change makes every status surface (CLI list/show, toggle UI, web dashboard hub, TUI gateway) agree on one effective-state resolver (`_plugin_status`), reports bundled backends/platforms as `enabled` by default, and stops representing filesystem model-providers as general toggles. Consolidation is a clear improvement, and the test coverage (unit + integration + toggle no-op preservation) is strong.

Non-blocking observations:

- **Toggle persistence semantics changed from "deny-all-unchecked" to "persist-only-changed".** Previously every unchecked row was written to the disabled-list on save, which pinned off-by-default plugins explicitly and protected against any future auto-enable. Now only rows the user actually changed are written; untouched standalone plugins stay absent from `plugins.disabled`. Today that's correct (standalones are opt-in), but it removes the old defense-in-depth. If a plugin's `kind` ever changes (e.g. `standalone` → `backend`) or a future auto-enable path appears, an untouched plugin would silently become active. Consider whether preserving explicit denials for *all* unchecked non-default-on rows is worth the config noise trade-off.
- **`_discover_all_plugins` no longer returns user model-providers.** This is a deliberate, documented behavior change, but it affects every consumer of `_discover_all_plugins` beyond the surfaces updated here (plugins_cmd, web_server, tui_gateway). Worth a quick grep for other callers that may still expect user model-provider entries (e.g. anything that renders "installed providers" from this list) so the dashboard `plugins.manage`/list doesn't silently drop them.
- **Success message counts effective selection, not config entries.** `"General plugins: {len(chosen)} enabled, {len(plugin_keys) - len(chosen)} disabled"` counts UI-checked rows; a default-on backend with no `plugins.enabled` entry counts as "enabled" and an untouched opt-in standalone counts as "disabled" (it's really "not enabled"). Cosmetic, but "disabled" for something never disabled may confuse.
- **Minor naming consistency:** the dir-path variable is `_d` in `cmd_toggle`, `_dir` in `cmd_list`, and `dir_path` in `cmd_show` — harmless, but unifying would make the repeated `_plugin_status` call sites read more uniformly.

No blocking issues found; explicit disable still wins over the default-on path, and the no-op toggle case is well tested.

PR #97902, opened 2026-08-29
feat(image-gen): configure Codex host model

Ratzzz33:
## Summary

- add image_gen.openai-codex.host_model for the Responses API host model
- default to gpt-5.5 while keeping the image tool model (gpt-image-2) and quality tier independent
- resolve profile-specific values through the real Hermes config loader
- document host model versus image model/tier

## Tests

- scripts/run_tests.sh tests/plugins/image_gen/test_openai_codex_provider.py: 33 passed
- real temporary default/profile HERMES_HOME/config.yaml coverage, including profile isolation and fallback behavior
- uv run ruff check on touched Python files
- git diff HEAD^ --check

Independent final review and evidence verifier passed.

### DIFF
--- a/plugins/image_gen/openai-codex/__init__.py
+++ b/plugins/image_gen/openai-codex/__init__.py
@@ -14,6 +14,11 @@ Selection precedence for the tier (first hit wins):
 3. ``image_gen.model`` in ``config.yaml`` (when it's one of our tier IDs)
 4. :data:`DEFAULT_MODEL` — ``gpt-image-2-medium``
 
+The Codex chat model that hosts the ``image_generation`` tool defaults to
+``gpt-5.5`` for compatibility. It can be selected independently with
+``image_gen.openai-codex.host_model``; this never changes the underlying
+``gpt-image-2`` image model or its quality tier.
+
 Output is saved as PNG under ``$HERMES_HOME/cache/images/``. Source images for
 image-to-image/editing are sent as Responses ``input_image`` content parts.
 """
@@ -175,8 +180,17 @@ def _resolve_model() -> Tuple[str, Dict[str, Any]]:
 
     if candidate is not None:
         return candidate, _MODELS[candidate]
-
     return DEFAULT_MODEL, _MODELS[DEFAULT_MODEL]
+
+
+def _resolve_host_model() -> str:
+    """Return the Codex chat model that hosts ``image_generation``."""
+    cfg = _load_image_gen_config()
+    sub = cfg.get("openai-codex") if isinstance(cfg.get("openai-codex"), dict) else {}
+    candidate = sub.get("host_model") if isinstance(sub, dict) else None
+    if isinstance(candidate, str) and candidate.strip():
+        return candidate.strip()
+    return _CODEX_CHAT_MODEL
 
 
 def _read_codex_access_token() -> Optional[str]:
@@ -317,7 +331,7 @@ def _build_responses_payload(
     if input_images:
         content.extend(input_images)
     return {
-        "model": _CODEX_CHAT_MODEL,
+        "model": _resolve_host_model(),
         "store": False,
         "instructions": _CODEX_INSTRUCTIONS,
         "input": [{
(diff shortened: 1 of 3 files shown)


Enough1122:

Reviewed PR #97902 (`feat(image-gen): configure Codex host model`).

The change decouples the Codex chat model that *hosts* the `image_generation` tool from the underlying `gpt-image-2` image model, with `image_gen.openai-codex.host_model` as an independent override. The test covers per-profile override (via `set_hermes_home_override`), whitespace stripping, and the fallback for invalid values — good coverage of the intended behavior.

Non-blocking observations:

- **Config re-read per payload build.** `_resolve_host_model()` calls `_load_image_gen_config()` every time `_build_responses_payload` runs. This matches the existing `_resolve_model()` pattern, so it's consistent, but a batch of N images re-parses config N times. If this ever shows up in profiling, the resolved value could be cached or resolved once per provider/generation call.
- **`host_model` is validated only for non-empty string.** An unknown/unsupported model string (e.g. a typo like `gpt-5.6-sol ` with different casing) is passed straight through to the Responses API — there's no allowlist. That's probably fine (the API returns a clear error for unknown models, and an allowlist would churn), but worth being aware that `host_model` trusts the operator more than the `model` tier selection does.
- Minor: the docstring claims the value "defaults to `gpt-5.5` for compatibility" while `_CODEX_CHAT_MODEL` remains the single source — consistent, just confirming the default is intentional.

No blocking issues. The fix cleanly keeps `tools[0]["model"]` pinned to `gpt-image-2` while making the host selectable.

magnus919:
I can reproduce the failure on current Hermes v0.21.0 (`03f3b092`) with a valid Codex-authenticated desktop session and the current `image_gen.provider: openai-codex` path.

The request fails before image generation with:

```text
Codex Responses API returned HTTP 404: The model `gpt-5.5` does not exist or you do not have access to it.
```

The installed provider still hardcodes `_CODEX_CHAT_MODEL = "gpt-5.5"`, while `gpt-image-2-high` correctly maps to the `gpt-image-2` image tool with `quality: high`. This makes the host-model override in this PR directly relevant.

One caution: the configured Hermes model `gpt-5.6-luna` is a local/provider alias and should not automatically be assumed valid as the ChatGPT Codex Responses host model. The implementation should either accept an explicitly supported native Codex host-model ID or provide a current supported default, with the image model remaining independently pinned to `gpt-image-2`.

Filed by Jasper (AI agent on behalf of Magnus Hedemark)

alertxsto:
i reproduced this on my account with Hermes 0.21.0 using the Codex OAuth route.

The setup itself looked normal:

- Codex OAuth was authenticated
- image provider was `openai-codex`
- image model was `gpt-image-2-high`
- the active chat model was `gpt-5.6-luna-900k`

I sent the same request shape used by the plugin and changed only the host model:

- `gpt-5.5` returned HTTP 404 with `model_not_found`
- `gpt-5.6-luna` returned HTTP 200 and successfully produced an image

so the image provider, auth, and `gpt-image-2-high` selection were not the problem. The confusing part is that the Codex model catalog still listed `gpt-5.5` during the test, which makes this look account or route dependent rather than a simple missing-model configuration.

The practical issue is still the hardcoded host model. when that model is unavailable for a particular account, there is no config option to select another working Codex model. Making `host_model` configurable would let affected users recover without changing the selected `gpt-image-2` quality tier.

i also verified the workaround by generating a valid PNG with `gpt-5.6-luna`.

gaoanze888:
@Ratzzz33 — nice fix. Confirming #105536 overlaps this and I've closed mine in favor of #97902.

One optional delta: layering an env-var override (`OPENAI_CODEX_CHAT_MODEL`) on top of `image_gen.openai-codex.host_model`, so the precedence is env → config → default. This mirrors `_resolve_model`/`resolve_static_model` in the other image_gen providers and helps accounts rotating the host model per-deployment without touching `config.yaml`. Happy to open a small follow-up against this branch if there's interest; otherwise feel free to ignore.

_AI-assisted: drafted with AI assistance, reviewed by me._

PR #97903, opened 2026-08-29
fix(profiles): scope gateway service cleanup past an ambient home override

### LINKED ISSUE #97897: profile delete leaves orphaned launchd service that respawns the deleted profile
# Bug: `hermes profile delete` leaves orphaned launchd service (plist) and command alias

**Component:** `hermes_cli/profiles.py` → `_cleanup_gateway_service()` (line 1701); `hermes_cli/gateway.py` → `get_hermes_home()` override resolution (`hermes_constants.py`)
**Platform:** macOS (launchd). Likely also affects Linux (systemd) — see "Fix".
**Severity:** Medium — a deleted profile resurrects itself via launchd `KeepAlive`, and (when two profiles share a bot token) silently hijacks a live messaging platform.
**Reproduced:** 2026-08-29, macOS 26.6.2, Hermes (installed tree under `~/.hermes/hermes-agent/`)

---

## Summary

Deleting a named profile does **not** reliably remove its installed launchd service definition
(`~/Library/LaunchAgents/ai.hermes.gateway-<profile>.plist`) nor its command alias
(`~/.local/bin/<profile>`). Because the launchd plist carries `KeepAlive` (login auto-start +
restart-on-crash), an orphaned plist causes the deleted profile's gateway to **respawn repeatedly**
after `hermes gateway stop`, and — if a second profile shares the same Discord/Telegram token — to
**fight for and win that token**, leaving the user's real profile in a reconnect-failure loop.

---

## Symptoms (observed)

1. `profiles/rnobot/` directory deleted previously, yet `--profile rnobot gateway run --replace`
   kept running (PID 3540).
2. Repeated `hermes gateway stop` (which also unloads the launchd job) did NOT stop the process from
   coming back — it respawned within seconds.
3. Main profile (default) gateway logged, in a loop:
   ```
   ERROR gateway.platforms.base: [Discord] Discord bot token already in use (PID 3540).
   Stop the other gateway first.
   ...
   WARNING gateway.run: Gateway started with no connected platforms — 1 platform(s) queued for retry:
   discord: Discord bot token already in use (PID 3540).
   ```
   (default = 14,762 messages; rnobot = 0 messages — the empty profile held the only live connection.)
4. Orphaned artifacts found on disk:
   - `~/Library/LaunchAgents/ai.hermes.gateway-rnobot.plist` (created Jun 15)
   - `~/.local/bin/rnobot` (`#!/bin/sh \n exec hermes -p rnobot "$@"`)

Manual cleanup that resolved it: `hermes gateway stop --profile rnobot`
→ `rm ~/Library/LaunchAgents/ai.hermes.gateway-rnobot.plist`
→ `rm ~/.local/bin/rnobot` → `hermes profile delete rnobot` (dir)
→ `hermes gateway restart --profile default` (reclaimed token, `✓ discord connected`).

---

## Root-cause analysis

`delete_profile()` (`profiles.py:1463`) does call the right cleanups:
- `_cleanup_gateway_service(canon, profile_dir)` (line 1530 / defined line 1701)
- `remove_wrapper_script(canon)` (line 1549, gated by `has_wrapper`)

The **macOS branch** of `_cleanup_gateway_service` (lines 1731-1739) is:
```python
elif _platform.system() == "Darwin":
    plist_path = get_launchd_plist_path()
    if plist_path.exists():
        subprocess.run(["launchctl", "unload", str(plist_path)],
                       capture_output=True, check=False, timeout=10)
        plist_path.unlink(missing_ok=True)
        print("✓ Launchd service removed")
```
and the whole body is wrapped in `try/except Exception as e: print(f"⚠ Service cleanup: {e}")`
(lines 1708-1741) — i.e. **any failure is swallowed and only printed**, so deletion proceeds with
the plist/alias left behind.

`plist_path` comes from `get_launchd_plist_path()` (`gateway.py:2481`), which derives the name from
`_profile_suffix()` (`gateway.py:1727`). That helper reads `get_hermes_home()`:
```python
home = get_hermes_home().resolve()
default = get_default_hermes_root().resolve()
...
profiles_root = (default / "profiles").resolve()
rel = home.relative_to(profiles_root)   # expects ~/.hermes/profiles/<name>
```
`_cleanup_gateway_service` relies on temporarily setting the env var to make that resolve correctly
(line 1707-1709):
```python
old_home = os.environ.get("HERMES_HOME")
os.environ["HERMES_HOME"] = str(profile_dir)
```

**But `get_hermes_home()` resolution order is `context-local override → env → default`**
(`hermes_constants.py`, `get_hermes_home()` line 106; `get_hermes_home_override()` line 37 uses a
`contextvars.ContextVar`). The dashboard/web delete path (`web_server.py:15611
delete_profile_endpoint`) sets a context-local home override via `set_hermes_home_override`. A
context-local override **shadows the env var**, so the "set env" trick in
`_cleanup_gateway_service` does **not** take effect in that process — `get_hermes_home()` returns the
caller's (default) home → `rel = home.relative_to(profiles_root)` raises `ValueError` → the
`try/except` prints `⚠ Service cleanup: ...` → **the profile's plist is never unlinked** (and
`launchctl unload` targets the wrong path). Same mechanism can leave the alias if the wrapper
detection runs against a different home view.

Two independent problems, both real:
1. **Env-var shadowing:** relying on `os.environ["HERMES_HOME"]` when a context-local override is
   active. `_profile_suffix` cannot be driven by env in that path.
2. **Silent partial failure:** `try/except` swallows every cleanup error and `delete_profile`
   unconditionally `rmtree`s the directory — so the command reports success even though the
   launchd job (KeepAlive) + alias remain and the profile will respawn.

---

## Why it hurts

- **Resurrection:** orphaned launchd `KeepAlive` brings the "deleted" profile back. Users who
  deleted a profile see it reappear and may loop on `stop` forever (the reported "deleted it but it
  keeps coming back").
- **Token hijack (severe):** if two profiles share a bot token, the respawned empty profile wins the
  single live connection and the user's real profile sits in a `token already in use` reconnect loop,
  dropping all inbound messages for the real account.

---

## Suggested fix

In `_cleanup_gateway_service` (`profiles.py:1701`):
1. **Use the documented override API, not env.** Scope to the target profile with
   `set_hermes_home_override(str(profile_dir))` / `reset_hermes_home_override(token)` around the
   `get_launchd_plist_path()` / `get_launchd_label()` calls, so it works regardless of any ambient
   context-local override. (The `finally` block should reset the override token, not just the env var.)
2. **Verify, don't assume.** After `launchctl bootout`/`unload` + `unlink`, check
   `plist_path.exists()`; if it still exists (or `launchctl list | grep label` is non-empty),
   surface a loud, non-swallowed warning naming the exact file, and **do not report the delete as
   fully successful** — print what remains and how to remove it.
3. Prefer the existing correct `launchd_uninstall()` (`gateway.py:4277`) which does `launchctl
   bootout <domain>/<label>` (handles unloaded jobs better than bare `unload`) + `unlink`. The
   delete path currently hand-rolls `launchctl unload` instead of reusing it.
4. **Same for the alias:** confirm `remove_wrapper_script` ran against the same home view; on
   failure, print the leftover path.
5. Consider: `profile delete` should verify the directory is *gone and stays gone* (a post-delete
   `sleep` + re-stat) so a live KeepAlive respawn is caught and reported, not silently re-created.

---

## Reproduction (macOS)

```bash
hermes profile create rnobot
hermes gateway install --profile rnobot        # installs ai.hermes.gateway-rnobot.plist + alias
hermes profile delete rnobot                   # via dashboard/web, or CLI in a process that has a
                                               # context-local home override set
# observe:
ls ~/Library/LaunchAgents/ai.hermes.gateway-rnobot.plist   # STILL PRESENT
ls ~/.local/bin/rnobot                                    # STILL PRESENT
launchctl list | grep rnobot                             # job still loaded
# stop it and watch it respawn:
hermes gateway stop --profile rnobot && sleep 5 && ps aux | grep "profile rnobot"
```

**Expected:** `profile delete` removes the plist + alias and the gateway does not respawn.
**Actual:** plist + alias remain; `KeepAlive` respawns the gateway; if a second profile shares a
token, it hijacks the connection.

---

## Notes for maintainers

- This matches the project rubric's "fix the whole bug class": the same env-vs-override shadowing
  likely affects any other call site that drives `get_hermes_home()`/`_profile_suffix()` by mutating
  `os.environ["HERMES_HOME"]` from a context that may hold a `set_hermes_home_override`.
- A regression test should run `delete_profile` for a named profile **with a context-local override
  active** (simulating the dashboard path) and assert the plist file is gone and `launchctl` no
  longer lists the label.

liuhao1024:
## What does this PR do?

`hermes profile delete` could leave the deleted profile's launchd service definition (`ai.hermes.gateway-<profile>.plist`) behind, so launchd `KeepAlive` kept respawning the "deleted" gateway — and when two profiles shared a bot token, the resurrected empty profile held the only live connection (#97897).

Root cause: `_cleanup_gateway_service()` scopes `_profile_suffix()` to the deleted profile by mutating `os.environ["HERMES_HOME"]`. But `get_hermes_home()` resolves the **context-local override (`set_hermes_home_override`) before the env var** (documented in `hermes_constants.py`). When the delete runs inside a context that already holds an override — the dashboard's delete endpoint (`web_routers/profiles.py`), or profile-scoped gateway contexts (`gateway/run.py` `_profile_runtime_scope`) — the env value is shadowed, so:

- `_profile_suffix()` derives the **default** (or the wrong profile's) service name when the ambient override points at the default root;
- the deleted profile's plist is never unlinked (the orphan), and worse, the shadowed resolution can `launchctl unload` + **unlink the default profile's plist** instead;
- the broad `try/except` only prints `⚠ Service cleanup: ...`, so the delete reports success either way.

The fix scopes the override layer to `profile_dir` alongside the env var (`set_hermes_home_override` / `reset_hermes_home_override` token pair) and restores both in the `finally` block — exactly the scoping pattern `get_launchd_plist_path()`/`get_service_name()` document. The env mutation is kept (restored unchanged) so downstream consumers that read the env directly (e.g. the Windows branch in flight) are unaffected. The Linux/systemd branch resolves through the same `_profile_suffix()`, so it benefits from the same fix.

Scope note: the command-alias half of the report is a different mechanism — `remove_wrapper_script()` works under `Path.home()/.local/bin`, which does not consult `get_hermes_home()` at all, so it is not affected by this shadowing. This PR fixes the service-definition half only.

## Related Issue

Fixes #97897

## Type of Change

## Changes Made

- `hermes_cli/profiles.py` — `_cleanup_gateway_service()`: wrap the existing env scoping with a `set_hermes_home_override(profile_dir)` / `reset_hermes_home_override(token)` pair so an ambient context-local override can no longer shadow the deleted profile during service-name resolution; both layers restored in `finally`.
- `tests/hermes_cli/test_profile_delete_service_scope.py` — new regression tests: (1) with an ambient override active (the dashboard delete scenario) the target plist is unlinked and `launchctl` receives the target plist path, while the default profile's plist is untouched; (2) plain no-override deletes still work; (3) a swallowed cleanup failure still restores the env/override scoping.

## How to Test

1. `python -m pytest tests/hermes_cli/test_profile_delete_service_scope.py -v` → 3 passed (Observed result: 3 passed locally).
2. Reverse check: revert only the `hermes_cli/profiles.py` hunk and rerun → the ambient-override test fails (target plist survives, default plist gets unlinked), the no-override test still passes, confirming the regression test pins the reported bug and that plain CLI behavior is unchanged.
3. `python -m pytest tests/hermes_cli/test_profiles.py tests/hermes_cli/test_deleted_profile_tombstone.py -q` → 72 passed, 2 skipped (no regressions in the profile-delete neighborhood).
4. Tests mock `subprocess.run` and redirect `_launchd_user_home()` into `tmp_path`, so nothing touches the real `~/Library/LaunchAgents` or runs `launchctl`.

### Code

### DIFF
--- a/hermes_cli/profiles.py
+++ b/hermes_cli/profiles.py
@@ -1945,8 +1945,21 @@ def _cleanup_gateway_service(name: str, profile_dir: Path) -> None:
     import platform as _platform
 
     # Derive service name for this profile
-    # Temporarily set HERMES_HOME so _profile_suffix resolves correctly
+    # Temporarily set HERMES_HOME so _profile_suffix resolves correctly.
+    # get_hermes_home() checks the context-local override BEFORE the env
+    # var, so mutating only os.environ leaves an ambient override (the
+    # dashboard delete endpoint, profile-scoped gateway contexts) shadowing
+    # the target profile: _profile_suffix() then derives the default — or
+    # the wrong profile's — service name, and the real plist/service file
+    # survives the delete. Scope both layers to profile_dir and restore
+    # both in the finally block.
+    from hermes_constants import (
+        reset_hermes_home_override,
+        set_hermes_home_override,
+    )
+
     old_home = os.environ.get("HERMES_HOME")
+    token = set_hermes_home_override(profile_dir)
     try:
         os.environ["HERMES_HOME"] = str(profile_dir)
         from hermes_cli.gateway import get_service_name, get_launchd_plist_path
@@ -1982,6 +1995,7 @@ def _cleanup_gateway_service(name: str, profile_dir: Path) -> None:
     except Exception as e:
         print(f"⚠ Service cleanup: {e}")
     finally:
+        reset_hermes_home_override(token)
         if old_home is not None:
             os.environ["HERMES_HOME"] = old_home
         elif "HERMES_HOME" in os.environ:
--- a/tests/hermes_cli/test_profile_delete_service_scope.py
+++ b/tests/hermes_cli/test_profile_delete_service_scope.py
@@ -1,0 +1,151 @@
+"""Service cleanup during profile delete must survive an ambient home override.
+
+``_cleanup_gateway_service`` scopes ``_profile_suffix()`` to the profile being
+deleted by setting ``HERMES_HOME``. But ``get_hermes_home()`` resolves the
+context-local override (``set_hermes_home_override``) BEFORE the env var, so
+when the delete runs inside a context that already holds an override — the
+dashboard's delete endpoint, or a profile-scoped gateway context — the env
+trick is shadowed and the plist path resolves to the default (or the wrong
+profile's) service. The orphaned plist then respawns the "deleted" profile
+via launchd ``KeepAlive`` (#97897). These tests pin the fix: the override
+layer is scoped alongside the env var, on both the happy path and the
+swallowed-exception path.
+"""
+
+from __future__ import annotations
+
+import os
+import platform
+import subprocess
+from pathlib import Path
+
+import pytest
+
+from hermes_cli import gateway as gateway_mod
+from hermes_cli import profiles as profiles_mod
+from hermes_constants import (
+    get_hermes_home,
+    get_hermes_home_override,
+    reset_hermes_home_override,
+    set_hermes_home_override,
+)
+
+
+@pytest.fixture()
+def profile_env(tmp_path, monkeypatch):
+    """Isolate profile paths from the user's real Hermes installation."""
+    monkeypatch.setattr(Path, "home", lambda: tmp_path)
+    default_home = tmp_path / ".hermes"
+    default_home.mkdir()
+    monkeypatch.setenv("HERMES_HOME", str(default_home))
+    return default_home
+
+
+@pytest.fixture()
+def darwin_service_branch(monkeypatch):
+    """Force the launchd branch regardless of the host OS.
+
+    The Linux/systemd branch shells out to systemctl and the Darwin branch
+    to launchctl; both subprocess calls are mocked in the tests below, but
+    the plist assertions only make sense on the Darwin branch.
+    """
+    monkeypatch.setattr(platform, "system", lambda: "Darwin")
+
+
+@pytest.fixture()
+def launch_agents(tmp_path, monkeypatch):
+    """Redirect launchd artifacts away from the real ~/Library/LaunchAgents.
+
+    ``get_launchd_plist_path()`` appends ``Library/LaunchAgents`` itself, so
+    the fake user home is the tmp root and the plist dir lives one level
+    below it.
+    """
+    monkeypatch.setattr(gateway_mod, "_launchd_user_home", lambda: tmp_path)
+    agents_dir = tmp_path / "Library" / "LaunchAgents"
+    agents_dir.mkdir(parents=True)
+    return agents_dir
+
+
+def test_cleanup_removes_target_plist_under_ambient_home_override(
+    profile_env, launch_agents, darwin_service_branch, monkeypatch
+):
+    """An ambient context-local override must not shadow the deleted profile.
+
+    Before the fix, ``get_hermes_home()`` returned the caller's override, so
+    ``_profile_suffix()`` derived the default profile's service name: the
+    target plist survived (KeepAlive respawn) and the default's plist was
+    unlinked instead.
+    """
+    profile_dir = profile_env / "profiles" / "rnobot"
+    profile_dir.mkdir(parents=True)
+    target_plist = launch_agents / "ai.hermes.gateway-rnobot.plist"
+    target_plist.write_text("rnobot service")
+    default_plist = launch_agents / "ai.hermes.gateway.plist"
+    default_plist.write_text("default service")
+
+    launchctl_calls: list[list[str]] = []
+    monkeypatch.setattr(
+        subprocess, "run", lambda cmd, **kwargs: launchctl_calls.append(list(cmd))
+    )
+
+    # Simulate the dashboard delete endpoint / profile-scoped gateway
+    # context: a context-local override pointing at the default home is
+    # active while the target profile's service is cleaned up.
+    token = set_hermes_home_override(profile_env)
+    try:
+        profiles_mod._cleanup_gateway_service("rnobot", profile_dir)
+    finally:
+        reset_hermes_home_override(token)
+
+    assert not target_plist.exists(), (
+        "the deleted profile's plist must be unlinked, or launchd KeepAlive "
+        "respawns it (#97897)"
+    )
+    assert default_plist.exists(), (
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed PR #97903 (`fix(profiles): scope gateway service cleanup past an ambient home override`).

The root cause is clearly identified: `get_hermes_home()` resolves the context-local override before `HERMES_HOME` env, so the old env-only trick was shadowed inside the dashboard delete endpoint / profile-scoped gateway contexts, causing `_profile_suffix()` to derive the wrong (or default) service name — leaving the real plist in place (launchd `KeepAlive` respawns the "deleted" profile, #97897). Scoping both layers and restoring both in `finally` is the correct fix, and the three tests (ambient override present, plain CLI path, exception path) pin the important scenarios including override/env restoration.

Non-blocking observations:

- **Nested-override assumption.** The fix relies on `set_hermes_home_override`/`reset_hermes_home_override` saving/restoring the prior override via the token, so the ambient override survives after the cleanup returns. The first test exercises exactly this (ambient override to `profile_env` remains active after `_cleanup_gateway_service` returns and the caller resets it), so the semantics are confirmed — just worth keeping in mind that any other code in the profile-delete path that computes a HERMES_HOME-derived path could be subject to the same shadowing.
- **Scope check on sibling cleanup.** The PR targets `_cleanup_gateway_service` specifically. If profile delete also removes other HERMES_HOME-derived artifacts (caches, launchd agent labels elsewhere in `profiles.py`), they could share the same shadowing hazard. A quick scan of the delete path for other `get_hermes_home()`/`_profile_suffix` consumers would confirm this is the only spot that needs the dual-layer scoping.

No blocking issues. Well-documented, minimal, and the regression tests directly reproduce the bug.

liuhao1024:
Thanks for the review — the notes match the intent: the token-based save/restore is exactly what keeps the ambient context intact across the scoped cleanup, and the fix boundary was drawn at the one cleanup site that derives a launchd service name.

PR #97904, opened 2026-08-29
fix(desktop): keep the latest turns in view when returning to a chat

jorgemdnt:
## What does this PR do?

Stops a conversation from jumping up to older turns when you switch away and back.

Tab round-trips keep the transcript mounted (`hot-hidden` → `visible`) without changing `sessionKey`, so the session-switch settle loop never re-arms. Hiding also clamps the DOM to the live tail, which can leave `scrollTop` at 0 on a still-scrollable tail.

Coming back then prepended older turns while treating that 0 as a chosen reading position — the thread landed on old messages instead of the latest. Reveal now unsets settled and pins to the bottom before the backfill. Hide pins the clamped tail to its latest line.

## Related Issue

N/A — reproduced from a user report. Adjacent to the first-paint / prepend-restore work in `ThreadMessageList` (the 11.5k px jump on session load); that path keys off `sessionKey` and does not run on a keep-alive tab return.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `apps/desktop/src/components/assistant-ui/thread/list.tsx`: pin the hidden tail to the bottom; on `hot-hidden` → `visible`, unset load-settled and pin before backfill prepends.
- `apps/desktop/src/components/assistant-ui/thread/list.test.ts`: cover reveal detection and the mid-load prepend anchor.

## How to Test

1. Open a long chat, scroll to the latest turn, switch to another conversation (tab or sidebar session that keeps the previous pane mounted), switch back. The latest turns should still be in view.
2. From `apps/desktop`: `npx vitest run --project ui src/components/assistant-ui/thread/list.test.ts` (31 passed).

### Code

- [ ] I've run `pytest tests/ -q` and all tests pass — N/A: Desktop TypeScript-only change
- [x] I've tested on my platform: macOS 15.6 arm64, Node.js 26.3.0

- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — scroll restoration is platform-neutral

## Screenshots / Logs

No visual chrome change — the thread should stay on the latest turns after a round-trip.

| Head | `npx vitest run --project ui src/components/assistant-ui/thread/list.test.ts` |
|---|---|
| `origin/main` `b1ff8722a5` | helpers absent (new tests fail to import) |
| This head `2291d22712` | 31 passed |

### DIFF
--- a/apps/desktop/src/components/assistant-ui/thread/list.test.ts
+++ b/apps/desktop/src/components/assistant-ui/thread/list.test.ts
@@ -14,7 +14,8 @@ import {
   shouldResettleTranscript,
   subscribeToThreadForeground,
   transcriptBackfillFrameCount,
-  transcriptPaneBudget
+  transcriptPaneBudget,
+  transcriptSettleAdvance
 } from './list'
 
 afterEach(() => {
@@ -352,5 +353,41 @@ describe('prependAnchorFromBottom', () => {
   it('preserves a settled reading position through a prepend', () => {
     expect(prependAnchorFromBottom(true, 800, 200)).toBe(600)
     expect(prependAnchorFromBottom(true, 800, 0)).toBe(800)
+  })
+})
+
+describe('transcriptSettleAdvance', () => {
+  const filled = {
+    clientHeight: 800,
+    frame: 4,
+    lastHeight: 12_000,
+    paneBudget: 600,
+    renderBudget: 600,
+    scrollHeight: 12_000,
+    stableFrames: 1
+  }
+
+  it('does not settle while the window is still 0-tall (app reopen / boot)', () => {
+    const next = transcriptSettleAdvance({
+      ...filled,
+      clientHeight: 0,
+      frame: 20,
+      lastHeight: 0,
+      scrollHeight: 0,
+      stableFrames: 8
+    })
+
+    expect(next.done).toBe(false)
+    expect(next.frame).toBe(20)
+  })
+
+  it('does not settle while first-paint backfill is still catching up', () => {
+    const next = transcriptSettleAdvance({ ...filled, renderBudget: 20, stableFrames: 2 })
+
+    expect(next.done).toBe(false)
+  })
+
+  it('settles once the pane is filled and height holds', () => {
+    expect(transcriptSettleAdvance(filled).done).toBe(true)
   })
 })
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed PR #97904 (`fix(desktop): keep the latest turns in view when returning to a chat`).

The fix addresses two related scroll regressions: the session-switch settle loop racing first paint during app reopen (0-height viewport, `scrollTop` left at 0), and tab round-trips (`hot-hidden` → `visible`) where the hide clamp leaves `scrollTop` 0 that a later prepend treats as a chosen reading position. Extracting `shouldResettleTranscript`, `prependAnchorFromBottom`, and `transcriptSettleAdvance` with unit tests is a good structure, and the tests cover the exact bug scenarios (4000px tail + 11kpx prepend; boot 0-height; backfill not caught up).

Non-blocking observations:

- **Render-phase ref writes.** `renderBudgetRef.current = renderBudget` / `paneBudgetRef.current = paneBudget` are assigned during render. This works in practice for reading "latest" in the rAF callback, but it's a render-phase mutation of a ref, which React's strict/concurrent guidance discourages (and it can theoretically be torn in concurrent mode). Moving the sync into a layout effect would be more idiomatic, though the current approach is common and low-risk.
- **Settle loop writes `scrollTop` every frame.** `node.scrollTop = node.scrollHeight` now runs on every rAF tick, including frames where `transcriptSettleAdvance` will report "not done" (0-height viewport, backfill behind). It's harmless (idempotent pin-to-bottom), but it does mean the loop keeps forcing scroll while waiting for the pane to fill — if a user tries to scroll up mid-load, the pin fights them until `done`. The old code only scrolled after stable frames; this is a behavior shift worth being aware of (probably the intended "pin until settled", just noting it's more aggressive).
- **Frame counter frozen while the viewport is 0-height.** Since `transcriptSettleAdvance` returns `frame` unchanged when `clientHeight < 2`, the 90-frame ceiling can't bound a persistent 0-height window. In practice rAF won't fire while the window is hidden and the effect is cleaned up on unmount, so this is acceptable — just confirming the cap is intentionally a "real frames with a real box" budget.

No blocking issues found; the regression tests directly pin the reported failures.

OutThisLife:
Superseded by main: same restore-consolidation family as #102248 (3ccad12c0a + 47f0cdf3a2). Closing as superseded; the issue is closed as fixed on main.

PR #97911, opened 2026-08-29
feat(web): mobile interrupt + message composer for dashboard chat

ChronodeAi:
## Problem

The dashboard's embedded Chat (the mobile gateway UI) has no way to interrupt a running turn or queue a message on touch devices. The TUI's interrupt (Esc) and its type-while-busy queue are keyboard-only, so phone users must summon the on-screen keyboard into the terminal — and the iOS software keyboard has no Esc key at all.

## Fix

A compact composer bar under the terminal, rendered on narrow layouts only:

- **Interrupt button** — sends the ESC byte as its own stdin write, hitting the TUI's actual interrupt binding (`useInputHandlers` — `key.escape` aborts the turn). Same path a physical Esc takes.
- **Message field + Send** — the text is written to the PTY exactly like typed input (burst then Return, same coalescing-safe timing as the existing `/copy` handler). While the agent is busy, the TUI composer queues the text and drains it when the turn ends — the same queue the desktop app's queued messages land in. No new server surface: pure UI over the existing PTY.
- Newlines are collapsed on send: each `\n` is a composer submit, so a multi-line burst would submit fragments as separate turns.

Desktop (>1024px) is unchanged — keyboard users keep the native bindings.

## Testing

- `tsc -b` and `vite build` clean; behavior mirrors the proven `handleCopyLast` PTY-write pattern already shipped in this file.
- Manually exercised against a live `hermes dashboard` backend from iPhone (Brave iOS): interrupt stops a running turn; a message sent mid-turn appears in the TUI queue and fires when the turn completes.

### DIFF
--- a/agent/transports/codex.py
+++ b/agent/transports/codex.py
@@ -7,8 +7,11 @@ streaming, or the _run_codex_stream() call path.
 
 import hashlib
 import json
+import logging
 import re
 from typing import Any, Dict, List, Optional
+
+logger = logging.getLogger(__name__)
 
 # Cron fires build session_id as ``cron_<job_id>_<YYYYMMDD_HHMMSS>`` (see
 # cron/scheduler.py). The trailing timestamp is per-fire noise; stripped so
@@ -272,7 +275,10 @@ def _profile_declared_efforts(
                 inferred_profile = get_provider_profile(inferred)
                 if inferred_profile is not None:
                     declared = inferred_profile.supported_reasoning_efforts(model)
-    except Exception:
+    except Exception as exc:
+        # Fail-open by design: a broken profile hook must never block the
+        # request — the transport falls back to its default vocabulary.
+        logger.debug("profile-declared efforts lookup failed: %s", exc)
         return None
     if declared is None:
         return None
--- a/plugins/model-providers/router/__init__.py
+++ b/plugins/model-providers/router/__init__.py
@@ -279,6 +279,10 @@ def _efforts_cache_only() -> Optional[dict[str, list[str]]]:
 def _warm_efforts_async() -> None:
     """Refresh the efforts cache in the background, at most once per process."""
     global _warm_started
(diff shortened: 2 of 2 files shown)


Enough1122:

Reviewed PR #97911 (`feat(web): mobile interrupt + message composer for dashboard chat`).

A sensible, narrowly-scoped feature: on narrow viewports, an interrupt button sends a raw ESC to the PTY (mirroring the TUI's interrupt binding) and a composer lets touch users type a message, queue it while the agent is busy, and submit it — without summoning the on-screen keyboard into the terminal. Aria-labels and the disabled-when-empty send button are good.

Non-blocking observations:

- **Draft cleared before the `\r` submit actually lands.** `handleComposerSend` sends the text, schedules `\r` 100ms later, then immediately clears `composerDraft`. If the WebSocket closes within that 100ms window (agent exit, reconnect), the text was transmitted but never submitted (no `\r`), and the user's draft is already gone. Consider clearing the draft only once the `\r` has been sent (e.g. inside the `setTimeout`, after the WS-open guard passes), or moving the clear into a `.then`/callback so a failed submit preserves the text.
- **The 100ms `\r` delay is only a server-side batching hedge.** WebSocket delivery is ordered, so the text and `\r` arrive in order; the delay exists so the two writes are read as separate frames on the server (letting the TUI composer enter "typing/queue" state before the submit). That's fine, but the fixed 100ms is an implicit timing assumption — if the server read loop ever batches across a different boundary, a very long paste could theoretically be split. Low risk; just noting the coupling.
- **Multi-line collapse is a deliberate trade-off.** `/\s*\n\s*/g → " "` means pasted multi-line text is flattened. Documented in the comment as intentional (avoids early submissions), so this is a UX note rather than a bug — if literal newlines are ever needed, Shift+Enter could be wired to a different action, but on a mobile composer single-line is reasonable.

No blocking issues.

PR #96939, opened 2026-08-28
Feat/add new model and provider

Contentment003111:
## What does this PR do?

Adds the new **Hunyuan `hy4-preview`** model (1M context) and a new **`tencent-tokenplan`** provider, and folds the two Tencent providers into a single grouped entry in the interactive provider pickers.

Previously Hermes only supported the `tencent-tokenhub` provider (OpenAI-compatible, `tokenhub.tencentmaas.com`) with `hy3-preview` as its aux/default model. This PR:
- Introduces `hy4-preview` with its correct **1,024,000-token** context window.
- Adds `tencent-tokenplan`, a second Tencent endpoint that speaks the **Anthropic Messages** protocol via `api.lkeap.cloud.tencent.com/plan/anthropic`.
- Groups `tencent-tokenhub` + `tencent-tokenplan` under one top-level row in the pickers (same display-only mechanism Kimi/MiniMax/xAI already use), so the provider list stays clean while both endpoints remain individually addressable.

## Related Issue

Fixes #

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `agent/model_metadata.py`
  - Add `"hy4-preview": 1024000` to `DEFAULT_CONTEXT_LENGTHS` (1M context window).
  - Register `api.lkeap.cloud.tencent.com` → `tencent-tokenplan` in `_URL_TO_PROVIDER`.
- `hermes_cli/auth.py`
  - Add `tencent-tokenplan` `ProviderConfig` to `PROVIDER_REGISTRY` (api_key auth, base URL `https://api.lkeap.cloud.tencent.com/plan/anthropic`, env vars `TOKENPLAN_API_KEY` / `TOKENPLAN_BASE_URL`).
  - Register `tokenplan` / `tencent-lkeap` aliases in `resolve_provider`.
- `hermes_cli/providers.py`
  - Add `tencent-tokenplan` `HermesOverlay` with `transport="anthropic_messages"`.
  - Add `tokenplan` / `tencent-lkeap` aliases and a `_LABEL_OVERRIDES` entry ("Tencent TokenPlan").
- `hermes_cli/models.py`
[...]

PR #93548, opened 2026-08-24
feat(providers): add Ramp Router (router.com) provider plugin

Neel49:
## What does this PR do?

Adds **Ramp Router** ([docs.router.com](https://docs.router.com)) as a provider — an OpenAI Responses-compatible gateway at `https://api.router.com/v1` that routes each request across upstream providers (OpenAI, Anthropic, xAI, Fireworks, …). Same shape as the recent Meta Model API integration: a self-contained plugin plus exact-hostname routing clauses.

- **Plugin** (`plugins/model-providers/router/`): `api_mode=codex_responses`, `RAMP_ROUTER_API_KEY` auth, live account-scoped catalog from `GET /v1/models` (no hardcoded models — vendor docs mandate runtime reads).
- **Host mandate** (`hermes_cli/providers.py` + `runtime_provider.py`): `api.router.com` → Responses API, Router's native wire (`/v1/chat/completions` is a minimal compatibility shim added Aug 2026; per-model reasoning validation and prompt caching live on Responses).
- **One new hook**: `ProviderProfile.supported_reasoning_efforts(model)`. Router validates `reasoning.effort` per model (400 on unsupported levels; 400 on any reasoning field for non-reasoning models — verified live), so the profile clamps from a cached copy of the catalog. Tri-state contract, cache-only hot path, disk-mirrored, resolved by provider name or endpoint host. Profiles that don't override it see zero behavior change.

## Related Issue

None — invited by the Nous team (Ramp ↔ Nous partnership channel).

## Type of Change

- [x] ✨ New feature (non-breaking change that adds functionality)

## How to Test

1. `pytest tests/hermes_cli/test_router_provider.py tests/agent/transports/test_router_codex_efforts.py -q` (hermetic, no network)
2. With a key from `https://app.router.com/keys`:
   `RAMP_ROUTER_API_KEY=… hermes -z "Reply with exactly one word: pong" --provider router -m gpt-5.4-mini`

Verified live against api.router.com: one-shot chat, streaming, tool loops + follow-up turns on OpenAI- and Fireworks-served models, encrypted-reasoning replay, and the effort clamp turning a would-be 400 (`xhigh` on o3) into a working request.

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)
[...]

PR #28253, opened 2026-05-18
feat(providers): add Nebius Token Factory provider

amrrs:
## What does this PR do?

Adds Nebius Token Factory as a first-class Hermes model provider using the existing bundled `ProviderProfile` plugin system.

This enables Nebius Token Factory to appear in `hermes setup` and `hermes model`, supports normal API-key configuration through Hermes provider env vars, fetches models from the OpenAI-compatible `/v1/models` endpoint, routes inference through the existing OpenAI-compatible chat-completions transport, and sends Nebius-compatible top-level `reasoning_effort` for reasoning-capable models.

## Related Issue

N/A

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Added `plugins/model-providers/nebius-token-factory/__init__.py` with a Nebius Token Factory `ProviderProfile` subclass.
- Added `plugins/model-providers/nebius-token-factory/plugin.yaml`.
- Added provider aliases: `nebius`, `nebius-tokenfactory`, `nebius-tf`, `token-factory`, and `tokenfactory`.
- Wired provider metadata into:
  - `hermes_cli/providers.py`
  - `hermes_cli/models.py`
  - `hermes_cli/model_normalize.py`
  - `agent/model_metadata.py`
- Added Nebius top-level `reasoning_effort` handling for reasoning-capable models.
- Added tests for provider wiring, model fetching fallback, reasoning emit/skip behavior, transport kwargs, and `nebius/<model>` prefix normalization.
- Updated provider discovery coverage with a Nebius spot-check while keeping the dynamic provider-count invariant.

## How to Test

1. Run the focused provider tests:

   ```bash
   uv run --extra dev pytest tests/hermes_cli/test_nebius_token_factory_provider.py tests/providers/test_plugin_discovery.py -q
   ```

2. Run lint on the touched files:

   ```bash
   uv run --extra dev ruff check plugins/model-providers/nebius-token-factory/__init__.py tests/hermes_cli/test_nebius_token_factory_provider.py hermes_cli/providers.py hermes_cli/models.py hermes_cli/model_normalize.py agent/model_metadata.py tests/providers/test_plugin_discovery.py
   ```

[...]

PR #97912, opened 2026-08-29
chore: map provider-salvage contributor emails (Neel49, amrrs)

kshitijk4poor:
Email→login mappings for the provider salvage PRs that follow: Neel49 (#93548 Ramp Router, both emails) and amrrs (#28253 Nebius Token Factory). simonweng@tencent.com (#96939) is already mapped. Mappings verified against each PR's commit author metadata.

### DIFF
--- a/contributors/emails/1littlecoder@gmail.com
+++ b/contributors/emails/1littlecoder@gmail.com
@@ -1,0 +1,2 @@
+amrrs
+# PR #28253 salvage (Nebius Token Factory provider)
--- a/contributors/emails/neel.patel@ramp.com
+++ b/contributors/emails/neel.patel@ramp.com
@@ -1,0 +1,2 @@
+Neel49
+# PR #93548 salvage (Ramp Router provider)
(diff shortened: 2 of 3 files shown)


Enough1122:

Reviewed PR #97912 (`chore: map provider-salvage contributor emails (Neel49, amrrs)`).

Three new contributor-email mapping files, matching the existing convention (first line = display name, second line = `#` provenance comment, cf. `0301chris@gmail.com`, `0xarkstar@users.noreply.github.com`). The local checkout shows no pre-existing `amrrs` or `Neel49` mappings and no conflicting files for these emails, so no collision risk from this change.

Non-blocking observations:

- **Two emails for one contributor (Neel49).** `neel.patel@ramp.com` (work/company email) and `neel49@users.noreply.github.com` (GitHub noreply) both map to `Neel49`. This is normal (a contributor using different author emails across commits), but worth a one-line sanity check that `neel.patel@ramp.com` is actually the same person — if the Ramp PR was authored by a different `neel.patel`, this would silently misattribute. The PR title ("salvage") implies it's the intent.
- **Dedup behavior.** If the email-mapping machinery treats a mapped email as "already attributed" and skips re-authorship, mapping both a company email *and* a noreply email for the same contributor could double-count one person as two authors. Verify the downstream tooling (whatever consumes `contributors/emails/`) handles multi-email-to-one-name without inflating the contributor count.

No blocking issues — purely additive metadata.

PR #49841, opened 2026-06-20
fix(desktop): raise remote liveness probe timeout from 2.5s to 10s

GratefulDave:
## Summary

Fixes the connect → immediately-stale → reconnect loop reported in #49787 for remote backends reached via high-latency paths (Tailscale relay, cross-region VPN).

## Root cause

`hermes:connection:revalidate` probes `/api/status` with a hardcoded `2_500`ms timeout:

```js
// main.cjs:5309 (before)
await fetchPublicJson(`${base}/api/status`, { timeoutMs: 2_500 })
```

This fires on every sleep/wake (`onPowerResume`), network online event, and `visibilitychange`. On a Tailscale-connected remote, relay latency can intermittently exceed 2.5s — causing the probe to time out and drop the cached connection as stale even when the backend is perfectly healthy. The result is an infinite loop:

```
Remote Hermes backend is ready
→ Cached remote Hermes backend failed liveness probe; dropping stale connection.
→ startHermes() rebuilds from scratch (boot progress overlay reappears at ~24%)
→ next wake/focus event fires revalidateConnection() again
→ same 2.5s race → drop again → loop
```

This is distinct from the bootstrap marker issue also reported in #49787 (setup screen on every launch), which has a separate fix.

## Fix

Raise the revalidate timeout to `10_000`ms, matching the timeout already used for the connection test (`fetchPublicJson` at line 4437):

```js
// main.cjs:5309 (after)
await fetchPublicJson(`${base}/api/status`, { timeoutMs: 10_000 })
```

A genuinely dead backend still fails within 10s. A live backend over a high-latency relay no longer gets false-evicted.

## Verification

- Affected setup: Mac M5, macOS Tahoe 26.5.1, Hermes Desktop v0.17.0, remote backend on Mac mini via Tailscale tunnel
- Tailscale round-trip: ~223ms average with occasional spikes above 2.5s
- `GET /api/status` returns HTTP 200 with valid JSON when probed directly; Let's Encrypt cert, no TLS issues
- With `timeoutMs: 2_500`: desktop.log shows repeated `Cached remote Hermes backend failed liveness probe` within seconds of each `Remote Hermes backend is ready` — confirmed via 8–10 loop iterations per session
- With `timeoutMs: 10_000`: probe succeeds reliably; loop stops

## Related

- Closes #49787
- Related: #43913 (different symptom, same desktop boot area — local mode Python 3.9 detection loop)
- Related: PR #43983 (missing bootstrap marker → setup screen; complements this fix)

PR #97914, opened 2026-08-29
fix(desktop): probe /api/health on pooled SSH dispatch, not /api/status

bennybuoy:
## Bug Description

Windows Desktop SSH (no-mux `ssh -N -L`) connects, spawns the remote dashboard, then immediately flaps:

```
Pooled remote backend "conn:<id>::<profile>" failed its dispatch probe (Timed out connecting to Hermes backend after 2500ms)
connection closed (no-mux tunnels killed)
```

SSH auth and remote `hermes` already succeeded. The client then kills the tunnel.

This is not #49841 (stale, still edits `main.cjs`; background liveness is already 10s).

## Root Cause

`ensureHealthyPooledRemoteBackendForDispatch` probes `/api/status` with a 2.5s budget. Bootstrap already used cheap `/api/health`. Cold `/api/status` through a new Windows/Tailscale forward is ~2.5–3s / ~6KB; `/api/health` is ~20–50ms / 52 bytes. A miss retires the pooled backend and tears down every no-mux tunnel, so the next click pays a full respawn.

The function comment already called this a cheap health probe.

## Fix

- Dispatch probe path: `/api/health`
- Timeout: 5s (`DEFAULT_HEALTH_PROBE_TIMEOUT_MS`), still shorter than background liveness (10s `/api/status`)
- Regression assertions on path + timeout

Shared Electron code (Linux/macOS included). Symptom is worst on Windows because OpenSSH has no ControlMaster.

## How to Verify

1. `cd apps/desktop && npx vitest run electron/remote-liveness.test.ts`
2. Windows Desktop → SSH gateway. After a cold spawn, dispatch should not log `after 2500ms` or immediately `no-mux tunnels killed`.
3. `GET /api/health` on the local forward should stay ~20–50ms.

## Test Plan

- [x] Added regression test for this bug
- [x] Existing `remote-liveness` tests still pass (22/22)
- [x] Manual verification on Windows 11 → Linux SSH remote: 0× `2500ms` after rebuild; live `/api/health` 18–33ms

## Risk Assessment

Low — dispatch liveness only. Background/post-resume still probes `/api/status` at 10s. Current backends (0.19+) expose `/api/health`; a missing route fails dispatch and reconnects, same as a dead tunnel.

### DIFF
--- a/apps/desktop/electron/remote-liveness.test.ts
+++ b/apps/desktop/electron/remote-liveness.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it, vi } from 'vitest'
 
 import {
   ensureHealthyPooledRemoteBackendForDispatch,
+  POOLED_REMOTE_DISPATCH_PROBE_PATH,
   POOLED_REMOTE_DISPATCH_PROBE_TIMEOUT_MS,
   REMOTE_LIVENESS_FAILURE_LIMIT,
   REMOTE_LIVENESS_FAILURE_WINDOW_MS,
@@ -292,9 +293,11 @@ describe('ensureHealthyPooledRemoteBackendForDispatch', () => {
       })
     ).resolves.toBe(replacement)
 
-    expect(probe).toHaveBeenCalledWith(stale, '/api/status', {
+    expect(probe).toHaveBeenCalledWith(stale, POOLED_REMOTE_DISPATCH_PROBE_PATH, {
       timeoutMs: POOLED_REMOTE_DISPATCH_PROBE_TIMEOUT_MS
     })
+    expect(POOLED_REMOTE_DISPATCH_PROBE_PATH).toBe('/api/health')
+    expect(POOLED_REMOTE_DISPATCH_PROBE_TIMEOUT_MS).toBe(5_000)
     expect(retire).toHaveBeenCalledOnce()
     expect(reconnect).toHaveBeenCalledOnce()
   })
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed PR #97914 (`fix(desktop): probe /api/health on pooled SSH dispatch, not /api/status`).

The change makes the pooled-dispatch readiness probe hit `/api/health` instead of `/api/status` (whose cold payload through a no-mux SSH forward routinely exceeds the old 2.5s budget and kills the tunnel), and raises the timeout to 5s to match the health-probe budget while staying under the 10s background liveness window. The test now asserts the exported path constant *and* the timeout value, so the intent is pinned.

Non-blocking observations:

- **Is `/api/health` auth-free on the pooled backend?** The dispatch probe runs against a forwarded endpoint; if that backend gates `/api/health` behind basic auth (401 when unauthenticated), the probe would misclassify a healthy backend as dead — the same failure class this PR is fixing, just moved. Worth confirming `/api/health` returns 200 without credentials on the remote backend (or that the probe can authenticate).
- **Timeout value vs. the referenced constant.** The comment says the 5s "match[es] the health-probe budget (`DEFAULT_HEALTH_PROBE_TIMEOUT_MS`)", but the literal `5_000` is hardcoded rather than imported from that constant. If `DEFAULT_HEALTH_PROBE_TIMEOUT_MS` ever changes, this value silently drifts from the stated intent. Referencing the constant directly (if importable without a cycle) would keep the comment honest.
- Minor: `POOLED_REMOTE_DISPATCH_PROBE_PATH` and the timeout are now separate exports — fine, but the `path` constant is only used once; a named constant is still a reasonable choice for the test's sake.

No blocking issues. The change is small, well-commented, and directly addresses a real latency failure mode.

teknium1:
Thanks @bennybuoy — Landed on main via #108112 (merge d1bd7b00df) — your `/api/health` dispatch-probe commit is on main under your authorship. The 5 s budget was replaced by the 10 s liveness budget (5 s is under the measured 6–8 s cold start). Thanks!

PR #87612, opened 2026-08-16
fix(state): WAL probe fallback, bare-SELECT locking, writable_schema finally, read-pool self-heal

wanliqin:
## Summary

Four independent SQLite/state-layer fixes, one commit each:

- **fix(state): report delete, not wal, for the indeterminate journal-mode probe** — when the WAL-reset-bug probe is indeterminate, returning `"wal"` wrongly enables the lock-free read pool; the conservative `"delete"` keeps reads serialized through `_lock`. Closes #86515
- **fix(state): route four bare-SELECT read paths through `_read_ctx`** — `get_compression_lock_holder`, the `clear_session_activity_labels` fast path, `get_handoff_state`, and `list_pending_handoffs` read on the shared write connection and can observe uncommitted data. Closes #86516
- **fix(state_search): reset writable_schema in a finally** — the legacy-FTS demote left `PRAGMA writable_schema=ON` latched on the long-lived write connection when the DELETE raised. Closes #86517
- **fix(state): destroy broken read-pool connections instead of returning them** — `_read_ctx` now mirrors the write side's `_reconnect_after_notadb` self-heal: a `DatabaseError` checkout is closed (permit released) instead of being returned to the pool. Closes #86518

## Tests

359 passed / 19 skipped across 15 related test files (hermes_state, session_db read pool/path split, state_db malformed/notadb self-heal, FTS search), plus 150 passed in `tests/hermes_state/`. No failures.

Issue #86516, opened 2026-08-14
state: four read paths bypass self._lock / _read_ctx and read on the shared write connection (dirty reads of uncommitted transactions)


wanliqin:
## Summary

Four read paths execute bare `self._conn.execute(SELECT ...)` on the instance-wide shared write connection instead of going through `_read_ctx` / `self._lock`:
- `hermes_state.py` ~5741 `get_compression_lock_holder`
- `hermes_state.py` ~5819 `clear_session_activity_labels` fast path
- `hermes_state.py` ~11492 `get_handoff_state`
- `hermes_state.py` ~11514 `list_pending_handoffs`

When another thread is inside `_execute_write`'s BEGIN IMMEDIATE transaction, these SELECTs **join that uncommitted transaction** and can read rows that later roll back. Concretely, `list_pending_handoffs` may see an in-flight `request_handoff`'s `handoff_state='pending'` and the watcher can advance a handoff that ultimately rolls back (later CAS catches some of this, but pollers see phantom state). In non-WAL (DELETE) mode these reads also collide with writers (`database is locked`); the handoff sites swallow that into `None`/`[]`, producing false "no handoff" negatives. Every other read path in the codebase uses `_read_ctx` or `with self._lock` — these four are clear deviations.

## Suggested fix

Route all four through `with self._read_ctx() as conn:` (it degrades to the `self._lock` path automatically when WAL is inactive).

Found auditing v0.20.1 (v2026.8.13).

wanliqin:
I'll take this one — fix in progress, PR to follow.

jwilson411:
Opened a smallest-complete fix: https://github.com/NousResearch/hermes-agent/pull/99454

Routes `get_compression_lock_holder`, `clear_session_activity_labels` (fast path), `get_handoff_state`, and `list_pending_handoffs` through `_read_ctx()` so they no longer join an in-flight writer transaction. AST + concurrent-writer regression tests included.

teknium1:
Fix for all four sites you inventoried (plus a fifth found by an extended AST sweep — `self._conn` passed to a helper) is up in https://github.com/NousResearch/hermes-agent/pull/99502, with structural gates so no new bare `self._conn` read can land. Your early inventory was the first complete map of this class — credited in the PR body. Will auto-close on merge.

PR #97920, opened 2026-08-29
fix(state): disable sqlite statement cache on SessionDB connections

SourceFelony:
## Summary

`SessionDB` connections can hard-crash the interpreter (native `SIGSEGV`, no
Python traceback) when a single `check_same_thread=False` sqlite3 connection is
executed on more than one thread at once. This disables CPython's per-connection
sqlite3 statement LRU cache on every tracked SessionDB connection via the
documented `cached_statements=0` workaround.

## Symptom / evidence

The gateway crashes natively under read-load peaks. macOS crash dumps show an
identical signature on a background worker thread:

```
_pysqlite_query_execute
  → bounded_lru_cache_wrapper
    → _PyDict_GetItem_KnownHash   ← segfault (reads address 0x8)
```

Faulting thread is `thread_run → pythread_wrapper` (a worker), never the main
loop. Runtime Python 3.11.15, sqlite3 3.53.1.

## Root cause

- The read path borrows pooled connections, but when the pool is at capacity
  (`_READ_POOL_MAX` reached), `_read_ctx` falls back to serving the read from
  the **shared writer connection** (`self._conn`), opened with
  `check_same_thread=False`.
- That means two threads — the dedicated writer plus whichever reader blew past
  the pool ceiling — can issue `.execute()` on the same `sqlite3.Connection` at
  once.
- CPython's `sqlite3` module keeps an internal per-connection LRU cache for
  prepared statements. Concurrent `.execute()` on one connection corrupts that
  cache dict; the interpreter walks a corrupted node and dies natively.

CPython issues #118172 and #146471 document this class of cache corruption under
concurrent use of a shared connection and recommend `cached_statements=0` as the
supported workaround.

## Fix

`cached_statements=0` is pinned at `_connect_tracked_db`, the choke point every
SessionDB connection flows through — including the pool fallback that borrows
the writer connection. This removes the corrupted-cache race vector for the
entire bug class rather than only widening the pool.

## Why this approach

- **One-line, low-risk, no behavior change** to SQL semantics or transaction
  handling — it only disables statement caching.
- **Covers every connection**, including the fallback path that was crashing,
  without depending on pool accounting.
- Disabling the statement cache on these connections costs prepared-statement
  re-preparation per execute, a negligible price for session-state queries.

## Testing

- New regression test `test_tracked_connections_disable_sqlite_statement_cache`
  proves every tracked connection gets `cached_statements=0`.
- Existing pool tests pass (concurrency-correctness paths remain green).
- Live verification: a real `_connect_tracked_db` connection was exercised from
  a second thread (the scenario that previously segfaulted) with no errors.

## Notes

- This is a genuine upstream gap: no existing issue/PR references
  `cached_statements` in `hermes_state.py`.
- Local reproductions were load-triggered; the crash signature above is from
  macOS `.ips` crash reports.

### DIFF
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -3893,6 +3893,12 @@ class SessionTurnLeaseLostError(RuntimeError):
 def _connect_tracked_db(path, tracking_path=None, **kwargs):
     """``sqlite3.connect`` that registers the open fd for lock-safety.
 
+    The CPython sqlite3 statement cache is disabled for every SessionDB
+    connection.  These connections may be handed across threads via
+    ``check_same_thread=False``; CPython issues #118172 and #146471 document
+    cache corruption and native crashes under concurrent use, with
+    ``cached_statements=0`` as the supported workaround.
+
     While a connection is live, byte-level probes of the same file are
     refused: an ``open()``/``close()`` cancels every POSIX advisory lock this
     process holds on it -- including a running VACUUM's EXCLUSIVE lock.
@@ -3904,6 +3910,7 @@ def _connect_tracked_db(path, tracking_path=None, **kwargs):
     connect would disable the guard for the lifetime of that connection,
     which is precisely the failure mode this module exists to prevent.
     """
+    kwargs.setdefault("cached_statements", 0)
     try:
         from hermes_cli.sqlite_safe_read import connect_tracked
     except ImportError:
--- a/tests/test_session_db_read_conn_pool.py
+++ b/tests/test_session_db_read_conn_pool.py
@@ -66,6 +66,36 @@ def _read(db):
     db.get_messages("s1")
 
 
+def test_tracked_connections_disable_sqlite_statement_cache(monkeypatch, tmp_path):
+    """Shared SessionDB connections must not use CPython's unsafe cache.
+
+    CPython issues #118172 and #146471 show that concurrent use of a
+    ``check_same_thread=False`` connection can corrupt the sqlite3 statement
+    cache and segfault in ``bounded_lru_cache_wrapper``.  Every SessionDB
+    connection passes through ``_connect_tracked_db``, so pin the documented
+    ``cached_statements=0`` workaround at that choke point.
+    """
+    import hermes_cli.sqlite_safe_read as safe_read
+    import hermes_state as state
+
+    observed = {}
+    sentinel = object()
+
+    def fake_connect_tracked(path, **kwargs):
+        observed.update(kwargs)
+        return sentinel
+
+    monkeypatch.setattr(safe_read, "connect_tracked", fake_connect_tracked)
+
+    result = state._connect_tracked_db(
+        tmp_path / "state.db",
+        check_same_thread=False,
+    )
+
+    assert result is sentinel
+    assert observed["cached_statements"] == 0
+
+
 @pytest.mark.requires_wal
 def test_read_pool_is_bounded_across_many_threads(db):
     """150 short-lived reader threads must not pin 150 connections.


Enough1122:

Reviewed PR #97920 (`fix(state): disable sqlite statement cache on SessionDB connections`).

The change pins `cached_statements=0` on every `_connect_tracked_db` connection, citing CPython issues #118172 and #146471 (statement-cache corruption / segfaults when a `check_same_thread=False` connection is used concurrently). That is the documented upstream workaround, and applying it at the single choke point every SessionDB connection passes through is the right location. The test cleanly verifies the kwarg reaches `connect_tracked` without needing a real SQLite/CPython version.

Non-blocking observations:

- **`setdefault` lets a caller re-enable the unsafe cache.** `kwargs.setdefault("cached_statements", 0)` respects an explicit non-zero value if any future caller passes one. Today no caller does, so this is fine — just worth remembering that the guard is a default, not a hard invariant. If the corruption is severe enough that no caller should be able to opt back in accidentally, a forced `kwargs["cached_statements"] = 0` would be stricter (at the cost of overriding an intentional override).
- **Performance cost of `cached_statements=0` is unmeasured.** Every prepared statement is recompiled per execution. For hot SessionDB read paths (the reader-pool is exercised across many threads) this can add measurable latency. Given the alternative is native crashes, it's clearly the right trade-off — but a quick benchmark on the reader pool would confirm the impact stays acceptable, and would be worth documenting if it shows up in profiling.
- Minor: the docstring and test both cite the CPython issue numbers, which is good for future archaeology.

No blocking issues. Focused, correct, well-referenced.

andrexibiza (commented):
Reviewed exact head `978ac97274ca96f0bf4d47cb93e1b9e50de25ebc` against base `e387cbc0aa0fc89560bc14438762c7722663db05`, current `main@3f36c87e1ebdfbf7d14a88229dc9be222c12ea89`, the full two-file diff, the existing review comment, current SessionDB read-pool/writer code, the read-pool tests and their history, CPython #118172/#146471, and the live state-layer interlocks below. The field crash signature is valuable, and putting a mitigation at `_connect_tracked_db` is mechanically clean. I cannot clear this head yet, because the causal path claimed by the PR is not the path the code actually executes.

### P1 — the stated overflow-reader race is already serialized; the live same-connection races are elsewhere

The PR says that when `_READ_POOL_MAX` is exhausted, an overflow reader and the dedicated writer can issue `.execute()` concurrently on `self._conn`. On this exact head, `_read_ctx()` does the opposite: after `_checkout_read_conn()` returns `None`, it executes the fallback as:

```python
with self._lock:
    yield self._conn
```

and `_execute_write()` acquires that same `self._lock` before `BEGIN IMMEDIATE` and the transaction callback. The read-pool architecture deliberately established this invariant in merged #83406 (the state-pool work there preserves Yishova's #76700 contribution): past the descriptor ceiling, readers *convoy on the locked writer connection*. `tests/test_session_db_read_conn_pool.py::test_exhausted_permits_fall_back_to_the_writer_connection` also pins exactly that topology.

So `check_same_thread=False` by itself is not enough to create the CPython failure shape here. It permits a connection to move across threads; the cited bugs require overlapping use of the same connection/cursor. The pool connections are exclusively checked out and returned, and the overflow writer fallback is lock-serialized.

There **are** still real same-writer-connection ownership violations on current main, but they are different call sites: `get_compression_lock_holder`, the no-op probe in `clear_session_activity_labels`, `get_handoff_state`, and `list_pending_handoffs` still execute bare `self._conn` reads outside `_read_ctx()` / `self._lock`. That exact class is already tracked by #86516 (wanliqin), was attempted in closed #87612, and has a current structural repair in #90734 (Kyzcreig), whose regression suite actually hammers those readers against turn-shaped writes. #93992 (alithnar-bot) is complementary: it fixes the `clear_session_activity_labels` surface plus a teardown-vs-worker lifetime race associated with the same macOS `_PyDict_GetItem_KnownHash -> bounded_lru_cache_wrapper -> _pysqlite_query_execute` crash signature.

That topology matters for merge semantics: this PR is, at most, **defense-in-depth adjacent to #86516/#90734/#93992**, not a replacement for their ownership/lifetime repairs. Disabling the statement cache can change the manifestation of an unsafe concurrent-use path without making the path safe. CPython #146471 itself demonstrates that distinction: with `cached_statements=0`, the concurrent same-cursor reproducer turns from a segfault into `ProgrammingError: Recursive use of cursors not allowed` rather than becoming valid concurrency.

**Required before landing:** identify and test the actual Hermes concurrency edge this change is intended to contain. If the target is the four current bare writer reads, either compose the structural `_read_ctx()` ownership fix (preserving the #86516 → #87612/#90734 lineage) or explicitly make this a defense-in-depth mitigation that does not claim to close that class. If the target is a different path, name that path and prove overlapping use of the same connection/cursor on current code. The overflow fallback cannot be used as that proof because it holds `self._lock`.

### P1 — the new test proves configuration, not the crash/fix contract, and the upstream-version evidence does not establish the reported 3.11.15 failure

`test_tracked_connections_disable_sqlite_statement_cache` monkeypatches `connect_tracked` and asserts the kwarg is `0`. That is a useful wiring assertion, but it will pass whether or not Hermes ever concurrently executes one connection, and it cannot fail with the native bug this PR claims to repair. Likewise, exercising a real connection from a *second* thread sequentially is not the cited failure shape; simultaneous use is the load-bearing condition.

The version evidence also needs tightening before using those CPython issues as causal proof for the field crash reported here. #118172's reporter and sqlite3 maintainer explicitly confirmed that reproducer on Python 3.12/3.13 and **not on earlier versions (tested back to 3.8)**; the PR's incident runtime is Python **3.11.15**. #146471 reproduces on 3.14/3.15 and shares one cursor concurrently. Those issues support “concurrent pysqlite use can be unsafe and cache disabling can mitigate some manifestations”; they do not, as currently cited, prove that this 3.11.15 Hermes crash came from the locked overflow fallback.

**Required proof:** add a RED→GREEN concurrency regression or isolated subprocess reproducer that exercises the real Hermes owner boundary, on a runtime relevant to the incident if possible, and fails before the change for the same class. A subprocess is appropriate for a native-crash regression so a successful failure cannot take down pytest itself. If the crash is too nondeterministic to pin, the structural invariant is still testable: no concurrent access to the shared writer connection outside its lock / no close while a worker owns the connection. #90734's AST ownership gate and concurrent-reader harness are directly relevant evidence rather than duplicate work.

One smaller point becomes load-bearing only if the cache policy is retained as a hard safety invariant: `kwargs.setdefault("cached_statements", 0)` allows any caller to opt back into the cache. If “every SessionDB connection has cache disabled” is the intended invariant, use an unconditional assignment and test an attempted non-zero override; otherwise describe it as a default, not a pin.

### Composition / receipts

- #83406 is the merged read-pool foundation; preserve Yishova/#76700 credit and its locked-fallback contract.
- #86516 is the still-open canonical report for the four bare shared-writer reads; #87612 is a closed, unmerged earlier attempt; #90734 is the current broader structural repair candidate; #93992 is complementary teardown + one-reader ownership work. None should be silently superseded by this one-line mitigation.
- Merged #97676 moved 39 *properly locked* pure reads onto `_read_ctx()` for WAL performance and explicitly left #90734's four unlocked readers untouched. That increases the importance of keeping the `_read_ctx()` ownership model precise rather than treating all cross-thread connections as concurrently shared.
- Current main has advanced from this PR's base to `3f36c87e...`, but no commit since `e387cbc...` touches either `hermes_state.py` or `tests/test_session_db_read_conn_pool.py`, so the code-level findings above still apply to the current composition.
- Exact-head CI `33255944488`, Docker `33255944080`, and Nix `33255944076` are all `action_required` with **zero jobs**, so this head has no executable upstream landing receipt yet. The local claims are development evidence, not exact-head CI proof.

This is worth pursuing—the native crash evidence points at a serious pysqlite ownership class—but the strongest version of the fix is to repair the owner boundary first and use cache disabling only where a real residual reproducer proves it buys safety. That preserves the good part of this contribution without masking the already-known concurrent-connection surfaces.

teknium1:
Closing in favor of the class fix in https://github.com/NousResearch/hermes-agent/pull/99502. Your statement-cache analysis was correct and is cited — the crash really is inside pysqlite's per-connection LRU — but `cached_statements=0` shrinks the native crash window while leaving concurrent use of one connection undefined behavior (and re-prepares every statement on the hot write path). The consolidated fix removes the sharing itself: the pool-exhaustion fallback you identified serves the read from the writer connection **under `self._lock`**, so writer and fallback readers can no longer overlap, and an AST gate keeps it that way. Thanks for pinning the CPython issue numbers.

PR #72714, opened 2026-07-27
fix(cli): honor canonical clarify timeout after config merge

izumi0uu:
## What does this PR do?

Fixes the classic CLI continuing to time out clarify prompts after 120 seconds when the user configures only `agent.clarify_timeout`.

#69774 centralized timeout resolution with the intended precedence of explicit legacy `clarify.timeout` -> canonical `agent.clarify_timeout` -> 3600. However, `load_cli_config()` still injected `clarify.timeout: 120` into its defaults before merging user config. The resolver therefore could not distinguish that synthetic default from an explicit legacy override and returned 120.

This removes only the synthetic CLI default. The shared resolver and its backward-compatible precedence remain unchanged, so a genuinely user-authored `clarify.timeout` still wins.

## Related Issue

Fixes #72688

Follow-up to #69774.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Remove the legacy 120-second clarify default from `cli.py`'s merged CLI defaults.
- Add loader-to-resolver integration tests for canonical configuration and explicit legacy override behavior.

## How to Test

1. Write `agent.clarify_timeout: 987` to an isolated `config.yaml` without a top-level `clarify.timeout`.
2. Pass the result of the real `load_cli_config()` through `resolve_clarify_timeout()`.
3. Verify it returns 987; before this patch the same test returned 120.
4. Add an explicit `clarify.timeout: 42` and verify the backward-compatible override still returns 42.

```text
Before fix:
tests/cli/test_cli_clarify_timeout_config.py -> 1 failed, 1 passed
(canonical 987 resolved to 120)

After fix:
49 passed - CLI integration + clarify gateway tests
21 passed - config drift, managed scope, and env expansion tests
64 passed - CLI initialization and ignore-user-config tests
ruff check -> passed
git diff --check -> passed
```

Commands:

```bash
.venv/bin/pytest -q tests/cli/test_cli_clarify_timeout_config.py tests/tools/test_clarify_gateway.py
[...]

PR #97925, opened 2026-08-29
fix: clarify/approval prompt timeouts silently ignored <=0 (unlimited)

lorgauss:
## Summary

Two related bugs in modal-prompt timeout handling (clarify + approval
confirm dialogs), both violating the `<=0` = unlimited convention that
`resolve_clarify_timeout()` already documents for `agent.clarify_timeout`.

### Bug 1 — clarify timeout silently shadowed by a stale default

`load_cli_config()`'s built-in defaults dict hardcodes:

```python
"clarify": {
    "timeout": 120,  # Seconds to wait for a clarify answer before auto-proceeding
},
```

`resolve_clarify_timeout()` checks the legacy top-level `clarify.timeout`
key first, falling back to the canonical `agent.clarify_timeout` only
when that's `None`. Since almost nobody sets the legacy key explicitly,
the merge in `load_cli_config()` never overwrites it — the hardcoded
`120` survives into `CLI_CONFIG` and is indistinguishable from a real
user override. Result: setting `agent.clarify_timeout: 0` for
"unlimited" in `config.yaml` has no effect. Every clarify prompt still
times out at 120s regardless.

This is a precedence bug, not a stale-process issue — restarting the
CLI/gateway does not help.

**Fix:** remove the hardcoded `clarify.timeout` default from
`load_cli_config()`. `resolve_clarify_timeout()` already falls back to
`agent.clarify_timeout`, then `3600`, correctly once the false override
is gone.

### Bug 2 — approval prompt never supported `<=0` (unlimited) at all

The approval/confirm-dialog path (dangerous-command confirmation) never
implemented the unlimited convention:

```python
self._approval_deadline = _time.monotonic() + timeout
```

An `approvals.timeout: 0` was read as "expire immediately", not
"unlimited" — auto-denying the pending command instead of waiting on the
user. Fixed to match how clarify already handles it: a `None` deadline
means unlimited, the poll loop skips the expiry check when the deadline
is `None`, and the countdown hint line hides itself instead of printing
a nonsensical negative number.

## Why bundled together

Same file, same root convention (`<=0` → unlimited), same user-facing
symptom (operator wandering away from the terminal getting auto-answered
or auto-denied instead of waited on indefinitely).

## Test plan

- `ast.parse` clean on the modified file.
- Verified by code trace against the existing, already-shipped clarify
  single-question implementation (the prior partial fix for approval
  timeouts was applied by hand and confirmed live in a running session
  prior to this PR).
- No `pytest` binary available in this environment to run
  `tests/cli/test_cli_approval_ui.py` directly — flagging for CI/reviewer
  to run that suite plus any `clarify_gateway`/`resolve_clarify_timeout`
  coverage.

## Risk / rollback

Small, additive-only diff (20 insertions, 9 deletions), one file. No
config schema changes, no new env vars, no new core tools. Reverting is
a straight `git revert` of this commit.

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -517,9 +517,11 @@ def load_cli_config() -> Dict[str, Any]:
 
             "skin": "default",
         },
-        "clarify": {
-            "timeout": 120,  # Seconds to wait for a clarify answer before auto-proceeding
-        },
+        # NOTE: no top-level "clarify" default here on purpose. resolve_clarify_timeout()
+        # treats clarify.timeout as a legacy override that wins over agent.clarify_timeout
+        # whenever it's not None — a hardcoded default here would silently shadow the
+        # canonical agent.clarify_timeout for every user who never set clarify.timeout
+        # explicitly (that includes <=0 "unlimited", which never got a chance to apply).
         "code_execution": {
             "timeout": 300,    # Max seconds a sandbox script can run before being killed (5 min)
             "max_tool_calls": 50,  # Max RPC tool calls per execution
@@ -16045,7 +16047,9 @@ class HermesCLI(CLIAgentSetupMixin, CLICommandsMixin, CLIBillingMixin):
                 "selected": 0,
                 "response_queue": response_queue,
             }
-            self._approval_deadline = _time.monotonic() + timeout
+            # <=0 means unlimited (never auto-deny while the user is still
+            # deciding) — same convention as agent.clarify_timeout.
+            self._approval_deadline = None if timeout <= 0 else _time.monotonic() + timeout
 
             # Modal prompt — paint immediately, bypassing the throttle/resize
             # guard. A throttled paint here can be silently dropped (250ms
@@ -16073,9 +16077,11 @@ class HermesCLI(CLIAgentSetupMixin, CLICommandsMixin, CLIBillingMixin):
                     )
                     return result
                 except queue.Empty:
-                    remaining = self._approval_deadline - _time.monotonic()
-                    if remaining <= 0:
-                        break
+                    # None deadline = unlimited: never auto-deny, just keep polling.
+                    if self._approval_deadline is not None:
+                        remaining = self._approval_deadline - _time.monotonic()
+                        if remaining <= 0:
+                            break
                     now = _time.monotonic()
                     if now - _last_countdown_refresh >= 1.0:
                         _last_countdown_refresh = now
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed PR #97925 (`fix: clarify/approval prompt timeouts silently ignored <=0 (unlimited)`).

The core change is coherent: removing the hardcoded top-level `clarify.timeout: 120` default (so it can't shadow `agent.clarify_timeout`, including the `<=0 = unlimited` convention), and making `_approval_deadline` `None` for `timeout <= 0` in `_approval_callback`, with the polling loop and the countdown hint handling `None`. The `None`-means-unlimited representation is a clean fit.

Non-blocking observations (some evidence-based from a scan of `_approval_deadline` consumers):

- **A sibling approval path in `hermes_cli/callbacks.py` is not updated.** `approval_callback()` there reads `approvals.timeout` (default 300) and does `cli._approval_deadline = _time.monotonic() + timeout` then `remaining = cli._approval_deadline - _time.monotonic(); if remaining <= 0: break`. Under the new `<=0 = unlimited` convention, a user setting `approvals.timeout: 0` through *this* path would compute a negative deadline and immediately auto-deny — exactly the behavior this PR is removing from `cli.py`. Per the repo's own rubric ("fix the whole bug class — sibling call paths included"), worth confirming whether `callbacks.py::approval_callback` is a live path and, if so, giving it the same `None`-deadline treatment.
- **`_approval_deadline` is now `float | int | None`, but other consumers/initializers still assume a number.** `self._approval_deadline = 0` initializers and the `remaining = ... - _time.monotonic()` sites in `cli.py` (other callbacks, e.g. the one at ~line 13699 in the current tree) and `callbacks.py` would do arithmetic on `None` if a `None` deadline ever reaches them. The PR correctly guards the two sites it touches; a type annotation on the field (`Optional[float]`) plus a sweep of the remaining readers would make the invariant explicit.
- **No test changes accompany the new `<=0` semantics.** The diff is purely `cli.py` changes — the behavior `timeout <= 0 → never auto-deny` (and the removed default) deserves at least one regression test (e.g. `_approval_deadline is None` when timeout is 0, and the polling loop not breaking on a `None` deadline). Consider adding it.

No blocking issues; the direction is right, just incomplete on sibling paths and coverage.

lorgauss:
Pushed a regression test (a977fb8) covering the missing-coverage point.

On the sibling-path point: I checked `hermes_cli/callbacks.py::approval_callback()` closely. It does have the same bug shape in isolation (unconditional `_time.monotonic() + timeout`, no `<=0` handling) — but tracing every caller in the tree, that function is not imported or invoked anywhere. Its neighbor in the same file, `clarify_callback`, is equally orphaned; only the third function in that file (`prompt_for_secret`) is actually used, by `cli.py`. So this looks like dead/legacy code rather than a live user-facing bug — happy to be corrected if I missed a dynamic import or entry point, but a grep/AST sweep of the whole repo turned up no call site. Given that, I have not patched it, to avoid landing a no-op change; flagging here in case it's worth a separate cleanup issue (delete vs. keep for a planned future wiring).

teknium1:
Thanks @lorgauss. This was resolved by #114875, which covers the same fix; #114875 — fix(clarify): CLI clarify prompt honours agent.clarify_timeout instead of a hidden 120 s cap (#113873, salvage #72714) — is now merged on `main` at `1b50849d273b`, closing issues #113873. Closing this PR in favour of the landed change; if you see a case it does not cover, please open a fresh issue with the repro and tag it.

PR #31126, opened 2026-05-23
Fix Telegram group command mention cleanup

DanielLSM:
## Problem

Telegram group commands addressed to the current bot commonly use `/command@bot args`. The Telegram adapter stripped its own bot mention together with following whitespace, so the command and first argument were glued together before dispatch.

Example:

```text
/steer@ExampleHelper_bot hello -> /steerhello
```

That makes the gateway see `steerhello` instead of dispatching `/steer hello`.

## Fix

When cleaning this bot's Telegram trigger text, replace this bot's mention with one space instead of an empty string. This preserves the command/argument boundary while still removing only the current bot mention; mentions for other bots are left intact.

## Validation

- `python -m py_compile gateway/platforms/telegram.py tests/gateway/test_telegram_group_gating.py`
- `python -m pytest tests/gateway/test_telegram_group_gating.py -k 'clean_bot_trigger_text_keeps_group_command_arg_separator or observed_group_context_uses_shared_source' -q`

PR #97931, opened 2026-08-29
fix(telegram): inject bot identity into channel_prompt for group messages

inmean:
## Summary

Fix a latent issue where Telegram group messages addressed to the bot as `@<bot> <text>` were silently dropped after the leading handle was stripped.

### The bug

When a user types `@myninamaxbot hello` in a Telegram group, the adapter strips `@myninamaxbot` from `event.text` via `_clean_bot_trigger_text` (line 9243) before dispatching to the agent loop. The agent receives just `hello` with no indication that the message was addressed to it, and stays silent because the system prompt never tells it what its own @-mention handle is.

This was masked because the only place that injects a bot-identity hint into `channel_prompt` is `_apply_telegram_group_observe_attribution`, which is gated by `observe_unmentioned_group_messages` (off by default). Most installs never enable that setting, so the identity hint never fires — and the agent never knows its own handle.

### The fix

Make the bot-identity hint **unconditional** for group messages:
- A new method `_telegram_bot_identity_channel_prompt()` returns a short block announcing the bot's @-username and ID, plus a reminder that stripped mentions were addressed to it.
- `_apply_telegram_group_observe_attribution` now injects this block for every group message (DMs are unaffected), regardless of the `observe_unmentioned_group_messages` setting.
- The existing observe-attribution behavior is preserved: when that setting is on AND the chat is in the observe allowlist, the observe block is appended after the identity hint.

### Diff stats

```
plugins/platforms/telegram/adapter.py              |  56 +++++--
tests/gateway/test_telegram_group_identity_hint.py | 173 +++++++++++++++++++++
2 files changed, 217 insertions(+), 12 deletions(-)
```

### Tests

Five regression tests in `test_telegram_group_identity_hint.py`:
- `test_identity_prompt_injected_when_observe_unmentioned_off` — the bug case
- `test_identity_prompt_injected_when_observe_unmentioned_on` — both blocks fire
- `test_identity_prompt_skipped_for_dm` — DMs unaffected
- `test_identity_prompt_includes_handle_hint` — explicit recovery language
- `test_mention_strip_does_not_lose_intent` — end-to-end strip + identity propagation

**Verification:**
- 3/5 tests **FAIL** against the unpatched adapter (proves they catch the bug)
- 5/5 pass with the patch
- 24 existing telegram tests still pass (no regressions)

### Reproduction (before the fix)

1. Add the bot to a Telegram group with Send Messages permission.
2. Type `@myninamaxbot hello` in the group.
3. The adapter's gate accepts the message (`require_mention=false` default, `exclusive_bot_mentions` falls through when cache is empty).
4. `_clean_bot_trigger_text` strips the leading handle, so the agent receives `hello` as `event.text` with empty `channel_prompt`.
5. The agent doesn't know `@myninamaxbot` is its own handle and stays silent.

### After the fix

Same scenario: the agent receives `hello` as `event.text` AND a `channel_prompt` block that says `Your @-mention name in this group is @myninamaxbot ... When the current message contains your @-mention, treat it as addressed to you even if the leading handle was stripped from the message text. Reply accordingly.` The model responds.

### Related work

This builds on the upstream username-cache work in `08130a26f` ("fix(telegram): follow @username renames and support non-bot handles") which addressed the gate-side stale-handle issue. This PR addresses the complementary prompt-side issue: even with a correct cache, the agent wouldn't recognize mentions without an explicit identity hint.

Also related: PR #3870 introduced `_should_process_message` and the original mention-gate logic.

### DIFF
--- a/plugins/platforms/telegram/adapter.py
+++ b/plugins/platforms/telegram/adapter.py
@@ -9300,27 +9300,45 @@ class TelegramAdapter(BasePlatformAdapter):
         )
 
     def _apply_telegram_group_observe_attribution(self, event: MessageEvent) -> MessageEvent:
-        """Align triggered group turns with observed-history attribution."""
-        if not self._telegram_observe_unmentioned_group_messages():
-            return event
+        """Align triggered group turns with observed-history attribution.
+
+        Always injects the bot identity hint into ``channel_prompt`` for
+        group messages, even when ``observe_unmentioned_group_messages`` is
+        off — the agent needs to know its own @-mention handle to recognize
+        that messages addressed to ``@<bot>`` are requests for it. Without
+        this hint the agent treats ``@<bot> test`` as an unknown third-party
+        mention and stays silent.
+        """
         raw_message = getattr(event, "raw_message", None)
         if not raw_message or not self._is_group_chat(raw_message):
             return event
+
+        # Always-on identity hint for group messages. The bot needs to know
+        # its own handle to interpret "@<bot> test" as addressed-to-it
+        # rather than as a third-party mention to ignore.
+        identity_prompt = self._telegram_bot_identity_channel_prompt()
+        channel_prompt = (
+            f"{event.channel_prompt}\n\n{identity_prompt}"
+            if event.channel_prompt
+            else identity_prompt
+        )
+
+        if not self._telegram_observe_unmentioned_group_messages():
+            return dataclasses.replace(event, channel_prompt=channel_prompt)
+
+        # Observed-history attribution path (unchanged).
         chat_id_str = str(getattr(getattr(raw_message, "chat", None), "id", ""))
         allowed = self._telegram_observe_allowed_chats()
         if not allowed or chat_id_str not in allowed:
-            return event
+            return dataclasses.replace(event, channel_prompt=channel_prompt)
         shared_source = self._telegram_group_observe_shared_source(event.source)
         observe_prompt = self._telegram_group_observe_channel_prompt()
-        channel_prompt = f"{event.channel_prompt}\n\n{observe_prompt}" if event.channel_prompt else observe_prompt
+        channel_prompt = (
+            f"{channel_prompt}\n\n{observe_prompt}"
+            if channel_prompt
+            else observe_prompt
+        )
         if event.message_type == MessageType.COMMAND:
-            # Commands must retain the original source (with user_id) so
-            # slash-access control (_check_slash_access) can identify the
-            # sender.  Replacing the source with an anonymised shared source
-            # (user_id=None) causes admin-only commands like /new to be
-            # denied even when the sender is an admin, because
-            # SlashAccessPolicy.is_admin(None) is always False.
-            # Still inject channel_prompt for group context.
             return dataclasses.replace(
                 event,
                 channel_prompt=channel_prompt,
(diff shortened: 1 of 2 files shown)


Enough1122:

Reviewed PR #97931 (`fix(telegram): inject bot identity into channel_prompt for group messages`).

The root cause is clear and the fix is well-targeted: `_clean_bot_trigger_text` strips the leading `@<bot>` handle, so with `observe_unmentioned_group_messages` off (the default) a group message `@<bot> test` reached the agent loop as bare `test` with no channel-prompt context about who the bot is — so the model treated `@<bot>` as an unknown third-party mention and stayed silent. Always injecting a stable identity block into `channel_prompt` for group messages fixes it. The test suite is thorough (off/on paths, DM exclusion, mention-strip end-to-end).

Non-blocking observations:

- **Prompt-cache stability is preserved but depends on the block being byte-stable.** The identity block is now injected into every group message's `channel_prompt`. Since it's derived from the (constant) bot username/id, it stays identical across turns of a conversation, so the cached prompt prefix is not invalidated — good. Worth a deliberate check that `_current_bot_username()` is deterministic per conversation and doesn't start as one value and later change to another (e.g. empty → resolved after `get_me`), which would change the block mid-conversation and break caching.
- **`"unknown"` fallbacks can mislead the model.** `username = self._current_bot_username() or "unknown"` and `bot_id = ... or "unknown"` mean that before the bot identity is resolved, the block asserts `@-mention name in this group=@unknown` / `user_id=unknown`. The model may then believe its handle literally is `@unknown` and answer wrong addresses. Consider skipping the identity block (or emitting a non-assertive "bot identity not yet resolved") when the username is unavailable, rather than asserting `@unknown`.
- **Slight prompt-bloat per group turn.** The block is a few dozen tokens on every group message. Negligible in practice, but if group chat volume is high it's worth knowing the cost is per-message.
- Minor: exposing the raw numeric `user_id={bot_id}` to the model is harmless but arguably noise; the @-username is the actionable part.

No blocking issues; the fix addresses a real silent-failure bug with solid coverage.

Glucksberg:
I reproduced the same multi-bot addressing problem and opened #103337 with a proposed fix plus handler-level regression tests. Linking it here because the identity-context part overlaps with this PR, and I would prefer we converge on one solution rather than ship duplicate hints.

The additional failure is loss of the original recipient positions: `@research_bot , @ops_bot are you both listening?` reaches the first agent as `, @ops_bot are you both listening?`. A generic identity hint helps, but retaining both handles also matters for messages assigning different tasks to each bot.

The proposed patch preserves group conversation text and captions, and uses the original Telegram entities to report whether the current message explicitly mentioned the receiving bot. Replies, wake words, and open-group admission report `no`; slash-command cleanup and DMs keep their current behavior. The context is composed in the shared event builder, outside the cached session system prompt.

The new parameterized regressions fail on the unmodified base (11 failed / 1 passed). With the patch, the canonical runner reports 659 Telegram tests passed, 0 failed, 2 skipped. These are adapter/handler tests, not a claim that LLM output or Telegram bot-to-bot delivery is deterministic.

If you or maintainers prefer folding this into #97931, feel free to reuse commit Glucksberg/hermes-agent@4029b41b08 (including the preservation tests); I can close #103337 once the combined fix is covered. The separate command-argument cleanup work in #31126 is deliberately left alone.

teknium1:
Thanks @inmean — this landed on main via #104253 (merge bc31e04d83e): same symptom; superseded, credited. Closing this PR as superseded; the credit is in that PR's commits/body. Campaign tracker: #104154.

PR #90002, opened 2026-08-19
fix(kanban): dispatcher self-review guard + deterministic bootstrap-failure classification

brunopradof:
## Problem

Two independent classes of false `gave_up` on the kanban dispatcher, both
diagnosed against a live production incident (board `uss-platform-staging`,
2026-08-19): a card whose work was already complete and correctly handed
off was reaped as `crashed` by `detect_crashed_workers` on subsequent
stale/doomed respawns, ticking `consecutive_failures` until the breaker
tripped `gave_up` — a false negative that hides real, delivered work.

### 1. Self-review respawn guard

`kanban_request_review(...)` called **without** an explicit `reviewer=`
deliberately leaves `tasks.assignee` naming the **implementer** (see
`request_review`'s docstring — this is intentional design, not a bug). The
review-lane dispatcher used to key spawn eligibility purely on
`status='review' AND assignee is spawnable`, so it re-spawned the **same**
profile onto its **own** just-completed card. That worker has no
legitimate transition left to make — the card is already correctly in
`review` with real work delivered — and either:
- re-issues `kanban_request_review`/`kanban_complete`, both correctly
  rejected with `"task is not in running/ready"`, then exits `rc=0` with
  the task still `running`, or
- exits `rc=0` outright without a terminal call.

Either way `detect_crashed_workers` reaps that as a clean-exit protocol
violation. Measured case: runs 5258 and 5260 both spawned within 68s of an
already-successful `kanban_request_review` on the same card.

**Fix:** the review dispatch loop now resolves the latest
`review_requested` event's `(implementer, reviewer)` payload before
calling `claim_review_task`. When `reviewer is None` and `implementer`
still equals the task's current `assignee`, the spawn is skipped (bucketed
as `skipped_self_review`, **not** a failure) and a one-time event + card
comment explain why — mirroring the existing `skipped_nonspawnable` /
`_emit_nonspawnable_alert` pattern.

`request_review` was also extended to persist reviewer provenance across a
`changes_requested -> re-submit` cycle: a re-review that omits `reviewer=`
now correctly reuses the **original** reviewer from the latest
`changes_requested` event instead of being misclassified as self-review
just because the second call didn't repeat `reviewer=`.

### 2. Deterministic bootstrap-failure classification

A worker spawned with an unresolvable `--skills`/profile name (e.g. a
review lane force-loading a skill that isn't linked on the target profile)
dies during CLI init with `Error: Unknown skill(s): ...` or
[...]

PR #95181, opened 2026-08-26
fix(kanban): stop reviewer-less review dispatch from self-looping

astraltrekkin:
## What does this PR do?

A Kanban worker that calls `kanban_request_review` without a distinct reviewer is re-claimed on every dispatch tick and re-run as its own reviewer. `review_dispatch` defaults to true, `request_review` leaves the implementer as assignee when `reviewer` is omitted, and the review-lane respawn guard skips the usual success/PR dedup. That loop is a clean lifecycle exit, so neither `consecutive_failures` nor the protocol-violation breaker counts it. In the reported incident that burned 89 runs across 4 cards until the model plan ran out.

This change parks reviewer-less (or implementer-as-reviewer) handoffs for a human instead of spawning the same profile, and auto-blocks after `kanban.repeated_self_review_limit` consecutive `review_requested` self-handoffs (default 3, 0 disables). Distinct-reviewer dispatch is unchanged.

## Related Issue

Fixes #95141

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `hermes_cli/kanban_db.py` - skip review-lane spawn when the latest `review_requested` event has no distinct reviewer; auto-block self-review streaks with a sticky `blocked` event plus `self_review_loop_detected`
- `hermes_cli/config_defaults.py` - add `kanban.repeated_self_review_limit` (default 3)
- `hermes_cli/kanban.py` - surface `skipped_self_review` on dispatch CLI/JSON
- `cli-config.yaml.example` - document the new review-dispatch defaults
- `tests/hermes_cli/test_kanban_review_lifecycle.py` - park, same-profile reviewer, breaker, and sticky-block coverage
- `website/docs/user-guide/features/kanban.md` - document park-for-human and the breaker

## How to Test

1. Create a task assigned to profile `P`, claim it, and call `request_review` with no `reviewer`. Confirm status is `review` and the event payload has `reviewer: null`.
2. Run `dispatch_once` (or wait one gateway tick). Confirm the task stays in `review` and is listed as skipped self-review, not spawned as `P`.
[...]

PR #97933, opened 2026-08-29
fix(kanban): guard first-review self-review when reviewer is omitted (#97910)

### LINKED ISSUE #97910: [Bug]: kanban_request_review without `reviewer` sends the card to review but keeps the implementer as assignee. the author is then dispatched to review their own work
## Summary

`reviewer` is optional on `kanban_request_review`. When it is omitted, the task
moves to the `review` column but its assignee is unchanged, so the review
dispatcher hands the card straight back to the profile that just implemented it.
The four-eyes property the review lane exists for is lost silently — no error,
no warning, and the event trail looks healthy at a glance.

## Reproduction

1. Create a task assigned to profile A.
2. Let A implement and call `kanban_request_review(summary=...)` **without**
   `reviewer`.
3. Run a dispatcher pass.

## Why this is worth guarding in code rather than in the prompt

The review lane's own contract already knows who the implementer is: `request_review`
writes `{"implementer": ...}` onto the event, and `request_changes` reads it back to
route rework. So the code has, at transition time, everything it needs to see that
implementer == prospective reviewer.

Note the asymmetry with the re-review path, which is already careful here: when
`reviewer` is omitted on a *re*-review, `request_review` refuses unless it can
recover reviewer provenance from the last `changes_requested` event
("re-review has no durable reviewer provenance ... pass reviewer= explicitly").
The first review has no equivalent guard.

### Actual Behavior

A is spawned again on the same card, claimed from `review`.

Event trail from a real run (profile `dev`, card t_4a177ae4):

    753  run=42  review_requested  {"implementer": "dev", "reviewer": null}
    754  run=43  claimed           {"source_status": "review"}     <- dev again
    755  run=43  spawned

`task_runs` confirms both runs ran under the same profile:

    42  dev       review_requested  297s
    43  dev       review_requested  209s     <- this was the "review"

In this run the model happened to notice and re-issued
`kanban_request_review(reviewer="reviewer")` on its second pass, which is how the
card eventually reached a real reviewer. That was luck, not a mechanism — the card
body named `reviewer="reviewer"` verbatim and the argument was dropped anyway.

### Affected Component

Other

N/A (CLI only)

### Debug Report

```shell
## Environment

Hermes Agent v0.20.5, Linux aarch64, local OpenAI-compatible backends,
`kanban.review_dispatch: true`, dispatched via repeated `hermes kanban dispatch`.
```

### Operating System

Ubuntu 24.04 (DGX Spark)

### Python Version

_No response_

### Hermes Version

_No response_

```shell

```

_No response_

## Suggested fix

Refuse — or at minimum warn — when the resolved reviewer equals the implementer:

* reject with a diagnostic in the style of the re-review branch, e.g.
  "review would be assigned to the implementer (dev); pass reviewer= explicitly
  or set kanban.allow_self_review", and/or
* emit a `self_review` warning on the `review_requested` event so it is visible
  in `hermes kanban show` and in the dispatcher log.

An opt-in escape hatch seems right: single-profile boards may legitimately want
self-review, but it should be a choice, not the default outcome of an omitted
optional argument.

shellybotmoyer:
## Summary

Fixes #97910.

When `reviewer` is omitted on a **first review** (`kanban_request_review` without a prior `changes_requested` cycle), the assignee stays as the implementer, silently routing the card back to the same profile that just built it. The four-eyes property the review lane exists for is lost — no error, no warning, and the event trail looks healthy at a glance.

## Root Cause

In `request_review()`, the re-review path already guards against missing reviewer provenance (refuses unless it can recover the reviewer from the latest `changes_requested` event). The first-review path has no equivalent guard: when `reviewer is None` and there's no prior `changes_requested` run, `reviewer` stays `None`, `assignee_sql` is empty, and the assignee is never updated — it remains the implementer.

## Fix

Added a four-eyes guard after reviewer resolution:

1. **`self_review_allowed()` config helper** — checks `kanban.allow_self_review` (default `False`), following the same pattern as `review_dispatch_enabled()`.

2. **Guard in `request_review()`** — computes `effective_reviewer = reviewer if reviewer is not None else implementer`. If `effective_reviewer == implementer` and self-review is not opted in, returns a diagnostic error:
   ```
   review would be assigned to the implementer (dev); pass reviewer= explicitly
   or set kanban.allow_self_review: true to opt in
   ```

3. **`self_review` flag on event** — when self-review IS allowed (opted in via config), the `review_requested` event records `"self_review": true` so it's visible in `hermes kanban show` and dispatcher logs.

4. **Opt-in escape hatch** — single-profile boards that legitimately want self-review can set `kanban.allow_self_review: true` in `config.yaml`.

## Test Plan

- [x] `test_first_review_without_reviewer_is_rejected` — verifies the guard fires
- [x] `test_first_review_with_explicit_different_reviewer_succeeds` — verifies normal path works
- [x] `test_explicit_self_review_is_rejected` — verifies passing `reviewer=` equal to implementer is also blocked
- [x] `test_rereview_without_reviewer_uses_provenance` — verifies re-review still recovers prior reviewer
- [x] All 22 existing review lifecycle tests pass (updated 2 to pass `reviewer=` explicitly)
- [x] All 26 tests pass together

## Note

PR #90002 from @brunopradof addresses the same root issue at the **dispatcher** level (preventing respawn of the same profile). This PR addresses it at the **transition** level (preventing the silent self-assignment in the first place). The approaches are complementary.

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -6597,6 +6597,25 @@ def request_review(
                     )
                 reviewer = prior_reviewer
         reviewer = _canonical_assignee(reviewer) if reviewer is not None else None
+        # Four-eyes guard: refuse first-review self-review unless opted in
+        # via ``kanban.allow_self_review``.  When ``reviewer`` is omitted on a
+        # first review the assignee stays as the implementer, silently routing
+        # the card back to the same profile that just built it (#97910).
+        effective_reviewer = reviewer if reviewer is not None else implementer
+        if (
+            effective_reviewer is not None
+            and effective_reviewer == implementer
+            and not self_review_allowed()
+        ):
+            return _ret(
+                False,
+                "review would be assigned to the implementer "
+                f"({implementer}); pass reviewer= explicitly or set "
+                "kanban.allow_self_review: true to opt in",
+            )
+        self_review_warning = (
+            effective_reviewer == implementer and self_review_allowed()
+        )
         assignee_sql = ", assignee = ?" if reviewer is not None else ""
         params: tuple[Any, ...]
         if expected_run_id is None:
@@ -6654,6 +6673,7 @@ def request_review(
                 "summary": event_summary or None,
                 "implementer": implementer,
                 "reviewer": reviewer,
+                "self_review": True if self_review_warning else None,
             },
             run_id=run_id,
         )
@@ -9619,6 +9639,24 @@ def review_dispatch_enabled() -> bool:
         )
     except Exception:
         return True
+
+
+def self_review_allowed() -> bool:
+    """Return whether a profile may review its own implementation work.
+
+    The default is **False** because the review lane exists to enforce a
+    four-eyes separation: when ``reviewer`` is omitted on a first review the
+    assignee stays as the implementer, silently routing the card back to the
+    same profile that just built it.  Single-profile boards that legitimately
+    want self-review can opt in with ``kanban.allow_self_review: true``.
+    """
+    try:
+        from hermes_cli.config import load_config
+        return bool(
+            (load_config() or {}).get("kanban", {}).get("allow_self_review", False)
+        )
+    except Exception:
+        return False
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


alt-glitch:

Related: #95181 and #92349 address the same reviewer-less self-review loop with different behavior (parking or automatic distinct-reviewer selection). This PR rejects the handoff by default with an explicit self-review opt-in; maintainers should choose the preferred lifecycle semantics.

PR #97934, opened 2026-08-29
feat(whatsapp): per-contact toolset overrides via contact_toolsets

docsmrh:
## Summary

`platform_toolsets.whatsapp` is platform-wide, so every allowlisted contact shares one toolset list. That makes `allow_from` effectively all-or-nothing: admitting a low-trust contact — a courier, a client, a receptionist — so the agent can answer their messages also hands them whatever the platform list carries.

On a default install that list is not small. Resolving it on a real config:

```python
>>> sorted(_get_platform_tools(cfg, 'whatsapp'))
['browser', 'clarify', 'code_execution', 'computer_use', 'cronjob', 'delegation',
 'file', 'image_gen', 'kanban', 'memory', 'session_search', 'skills',
 'terminal', 'todo', 'tts', 'vision', 'web']
```

So an allowlisted WhatsApp DM reaches `terminal`, `file`, `code_execution` and `computer_use`. The only way to narrow that today is to narrow it for *every* WhatsApp contact, the operator included — which is why I hit this: I wanted a contact gatekept to "can talk, cannot touch the shell" and there was no way to express it.

## Approach

Implement `toolsets_for_source()` on `WhatsAppBehaviorMixin`, reading an optional `platforms.whatsapp.extra.contact_toolsets` map:

```yaml
platforms:
  whatsapp:
    extra:
      contact_toolsets:
        "923001234567": [web, vision, clarify]
        "923009876543@s.whatsapp.net": [web]
```

This deliberately adds **no new mechanism**. `toolsets_for_source()` is the existing per-source override hook that `WebhookAdapter` already uses for per-route toolsets, and the gateway validates whatever it returns through the same `_get_platform_tools` path as platform config. An entry therefore cannot self-grant a toolset that platform config could not: unknown and platform-restricted names are dropped rather than trusted (asserted in `test_override_validated_like_platform_config`).

Landing it on the shared mixin rather than the bridge adapter means the Baileys adapter and `WhatsAppCloudAdapter` both get it from one implementation.

## Design notes

- **Alias-form matching.** Keys go through the existing `_matches_whatsapp_allowlist`, so phone / JID / LID forms all resolve to one contact. WhatsApp delivers inbound senders in LID form while operators configure phone numbers — the reason raw set membership was already insufficient for `allow_from`.
- **Groups.** Group chats match the group JID first, then fall back to the sender, so a per-contact restriction still applies to that contact's messages inside a group. A group-wide entry wins over a per-sender one.
- **An empty list is refused, with a warning.** This is the sharp edge. The gateway's gate is `if override and isinstance(override, list)`, so `[]` is falsy and would fall through to the *platform* list — silently **widening** access, the exact opposite of what someone writing `contact_toolsets: {"923...": []}` intends. Honoring it as "no tools" would need a gateway-side sentinel change; refusing it loudly keeps this PR to one layer and cannot fail open. The warning names the alternative (remove them from the allowlist to deny entirely).
- **`"*"` is not special-cased.** Matching every contact is what `platform_toolsets.whatsapp` already does; honoring a wildcard here would shadow that setting.
- **No new env var.** Per `AGENTS.md`, this is behavioral config and lives in `config.yaml` only. `extra` is pass-through, so no whitelist needed.
- **Inert when unset.** Absent `contact_toolsets`, resolution is byte-identical to before — the method returns `None` and the platform path runs untouched.

## Test plan

23 new cases in `tests/gateway/test_whatsapp_contact_toolsets.py`, mirroring the structure of the existing `test_webhook_route_toolsets.py`:

- alias-form matching in both directions (bare number key ↔ JID sender, and the reverse)
- group JID entry, group→sender fallback, group-entry precedence
- malformed entries: non-list value, non-dict mapping, blank key, blank/whitespace toolset strings
- the empty-list refusal and its warning
- the wildcard non-match
- gateway integration: override replaces (not merges) the platform list, `terminal` is absent from the restricted result while present in the platform list, validation parity with platform config, config non-mutation, and adapter-exception fallback

Verified with `scripts/run_tests.sh` — **119 passed, 0 failed, 2 skipped** across the 9 related gateway suites:

```
tests/gateway/test_whatsapp_contact_toolsets.py          23✓   (new)
tests/gateway/test_whatsapp_cloud.py                     43✓
tests/gateway/test_webhook_route_toolsets.py             12✓
tests/gateway/test_whatsapp_connect.py                   10✓
tests/gateway/test_multiplex_toolsets_profile_isolation.py 9✓
tests/gateway/test_75349_whatsapp_multiplex_secret_scope.py 6✓
tests/gateway/test_whatsapp_cloud_allowed_users.py        6✓
tests/gateway/test_whatsapp_allowlist_lid_resolution.py   5✓
tests/gateway/test_api_server_toolset.py                  5✓
```

One honest note on my environment: `test_whatsapp_cloud.py` initially showed 10 failures (`'NoneType' object has no attribute 'Response'`). I stashed my change and re-ran on pristine `main` — **identical 10 failures**, so they were `aiohttp` missing from my venv, not a regression. Installing `aiohttp` cleared all 10. Flagging it in case CI images differ.

## What I did not do

- Did not touch the gateway's override gate, so "deny all tools" is still inexpressible. That needs a sentinel (`False` vs `[]`) and belongs in its own PR if wanted.
- Did not add per-contact toolsets to other platforms. The same gap exists for Telegram/Discord `allow_from`; this PR fixes the one I could test end-to-end rather than speculating across adapters.

### DIFF
--- a/gateway/platforms/whatsapp_common.py
+++ b/gateway/platforms/whatsapp_common.py
@@ -4,6 +4,7 @@ and the official WhatsApp Cloud API adapter.
 
 The mixin provides:
 - Allow-list / DM / group gating
+- Per-contact toolset overrides (``contact_toolsets``)
 - Mention detection (explicit @-mentions + configurable regex patterns)
 - Quoted-reply-to-bot detection
 - Broadcast / Channel / Newsletter filtering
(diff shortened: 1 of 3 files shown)


Enough1122:

### PR #97934 — feat(whatsapp): per-contact toolset overrides via contact_toolsets

**Summary:** Adds `WhatsAppBehaviorMixin.toolsets_for_source` allowing `platforms.whatsapp.extra.contact_toolsets` to map phone/JID/LID identifiers to per-contact toolset lists. Group JID is tried first then sender fallback; matching reuses `_matches_whatsapp_allowlist` so alias forms resolve. Empty or non-list entries are refused with a warning, wildcard `*` is intentionally ignored, and gateway integration validates overrides via `_get_platform_tools` so overrides cannot self-grant restricted toolsets.

- First-match wins in nested `candidates × mapping.items()` loop (`gateway/platforms/whatsapp_common.py:74`). Dict iteration order is stable in 3.7+, but if config YAML preserves insertion order the precedence may be surprising — consider documenting that key order matters for groups vs sender.
- Complexity is O(C·M); M is expected small (operator allowlist), so fine. No cap needed.
- `str(t).strip()` stringifies toolset names (`:90`); non-string entries become `"123"` and pass `cleaned` check, but downstream `_get_platform_tools` will drop unknown names anyway — harmless.
- Group-wide entry winning over sender (`test_group_entry_wins_over_sender_entry`) is intentional and documented, but operators expecting per-contact restriction to override group allowance may be surprised — docstring clarifies.

**Verdict:** Sound least-privilege narrowing for WhatsApp. Tests cover alias forms, group fallback, malformed entries, and gateway integration (override replaces platform list, validated, config not mutated). No blocking issues.

PR #97945, opened 2026-08-29
fix(desktop): allow Skills Hub links and clipboard writes

### LINKED ISSUE #91612: [Bug]: embedded Skills Hub sandbox blocks its own external links and clipboard controls
## Bug description

The Skills Hub embedded in Hermes Desktop renders controls that attempt to open external links and copy text, but the host iframe policy blocks those operations. In one normal browsing period, Electron recorded:

- 14 blocked popup attempts;
- 4 blocked Clipboard API calls.

The affected controls therefore appear clickable but do nothing.

## Root cause

`apps/desktop/src/app/skills/embedded-hub-picker.tsx` embeds the cross-origin hub with:

```tsx
<iframe
  sandbox="allow-scripts allow-same-origin"
  src={HUB_PICKER_URL}
/>
```

The sandbox omits popup/navigation capability, and the iframe has no clipboard permissions policy. The embedded page still exposes documentation/GitHub/Discord links and copy controls that depend on those capabilities.

## Expected behavior

User-visible controls in the embedded hub should either work or be replaced/disabled with an explicit explanation.

Security-preserving options:

1. Intercept approved external-link intents through a narrow `postMessage` contract and open them with the Desktop shell's safe external-browser capability.
2. Bridge copy actions through a narrow host clipboard capability.
3. If direct iframe capabilities are acceptable, add only the minimum sandbox/permissions policy (`allow-popups` / carefully considered `allow-popups-to-escape-sandbox`, plus scoped clipboard permissions).

A test should verify that an approved external link reaches the safe shell opener and a copy action reaches the clipboard without granting general navigation or Node access to the iframe.

## Environment
- OS: Windows 11
- Hermes Agent/Desktop: v0.20.4 (2026.8.18)
- Reported upstream: `fcbd1076`
- Local checkout: `fb27614addac115d55299bc6538ae112fd01f688`
- Install method: git
- Python: 3.11.16

sylvesterkaczmarek:
## What does this PR do?

Restores the embedded Skills Hub's user-visible external links and Copy controls while keeping the iframe policy narrow.

The picker currently allows only scripts and same-origin behavior. That blocks user-initiated popups from external links and Clipboard API writes used by the Copy control. This delegates only `clipboard-write` and adds popup capability. Clipboard read, top-level navigation, forms, downloads, and Node access remain unavailable.

## Related Issue

Fixes #91612

## Type of Change

## Changes Made

- `apps/desktop/src/app/skills/embedded-hub-picker.tsx`: allow clipboard writes and user-initiated popup links for the cross-origin Skills Hub iframe.
- `apps/desktop/src/app/skills/embedded-hub-picker.test.tsx`: pin the exact iframe permission and sandbox policy.

## How to Test

1. `cd apps/desktop`
2. Run `npx vitest run --project ui src/app/skills/embedded-hub-picker.test.tsx`
3. Open the embedded Skills Hub in Desktop, verify an external documentation/GitHub link opens, and verify Copy install command writes to the clipboard.

## Checklist

- [x] I've added regression coverage
- [x] I've considered the security and cross-platform impact
- [x] Documentation/config/tool-schema changes are N/A

### DIFF
--- a/apps/desktop/src/app/skills/embedded-hub-picker.test.tsx
+++ b/apps/desktop/src/app/skills/embedded-hub-picker.test.tsx
@@ -1,0 +1,73 @@
+// @vitest-environment jsdom
+import { render } from '@testing-library/react'
+import { describe, expect, it, vi } from 'vitest'
+
+vi.mock('@nanostores/react', () => ({
+  useStore: () => undefined
+}))
+
+vi.mock('@/components/ui/button', () => ({
+  Button: ({ children, size: _size, variant: _variant, ...props }: any) => <button {...props}>{children}</button>
+}))
+
+vi.mock('@/i18n', () => ({
+  useI18n: () => ({
+    t: {
+      skills: {
+        hub: {
+          actionFailed: 'Action failed',
+          actionLog: 'Action log',
+          alreadyInstalled: (name: string) => `${name} is already installed`,
+          installStarted: (name: string) => `Installing ${name}`,
+          pickerBrowse: 'Browse',
+          pickerHide: 'Hide',
+          pickerHint: 'Browse skills',
+          pickerTitle: 'Skills Hub',
+          updateAll: 'Update all',
+          updateStarted: 'Update started',
+          updating: 'Updating'
+        }
+      }
+    }
+  })
+}))
+
+vi.mock('@/lib/icons', () => ({
+  Loader2: () => <span />
+}))
+
+vi.mock('@/lib/use-session-slice', () => ({
+  useStoreSelector: () => false
+}))
+
+vi.mock('@/store/hub-actions', () => ({
+  $hubActions: {},
+  installHubSkill: vi.fn(),
+  UPDATE_ALL_KEY: 'update-all',
+  updateHubSkills: vi.fn()
+}))
+
+vi.mock('@/store/notifications', () => ({
+  notify: vi.fn(),
+  notifyError: vi.fn()
+}))
+
+vi.mock('@/store/panes', () => ({
+  $paneHeightOverride: () => ({}),
+  setPaneHeightOverride: vi.fn()
+}))
+
+import { EmbeddedHubPicker } from './embedded-hub-picker'
+
+describe('EmbeddedHubPicker iframe policy', () => {
+  it('allows user-initiated hub links and clipboard writes without broader iframe privileges', () => {
+    const { container } = render(<EmbeddedHubPicker installedNames={new Set()} />)
+    const iframe = container.querySelector('iframe')
+
+    expect(iframe).not.toBeNull()
+    expect(iframe?.getAttribute('allow')).toBe('clipboard-write')
+    expect(iframe?.getAttribute('sandbox')).toBe(
+      'allow-scripts allow-same-origin allow-popups allow-popups-to-escape-sandbox'
+    )
+  })
+})
--- a/apps/desktop/src/app/skills/embedded-hub-picker.test.tsx
+++ b/apps/desktop/src/app/skills/embedded-hub-picker.test.tsx
@@ -70,7 +70,7 @@ import { EmbeddedHubPicker } from './embedded-hub-picker'
 
 describe('EmbeddedHubPicker iframe policy', () => {
   it('allows user-initiated hub links and clipboard writes without broader iframe privileges', () => {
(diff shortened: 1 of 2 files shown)


Enough1122:

### PR #97945 — fix(desktop): allow Skills Hub links and clipboard writes

**Summary:** `EmbeddedHubPicker` iframe adds `allow="clipboard-write"` and expands sandbox to `allow-popups allow-popups-to-escape-sandbox` so hub links open externally and copy actions work without broadening iframe privileges. Test pins the exact `allow`/`sandbox` attributes.

- `allow-popups-to-escape-sandbox` lets opened pages escape the sandbox — intended for external hub links, but hub-controlled content could open arbitrary URLs. The trade-off is reasonable for user-initiated navigation; no `allow-top-navigation` remains, so top-window navigation is still blocked.
- `clipboard-write` is minimal least-privilege for copy buttons; no broader `allow` permissions added.

**Verdict:** Minimal, correctly scoped iframe policy change.

J80-droid:
This iframe change is necessary, but on current Desktop (`471ef5f`) it is not sufficient.

Hub `target="_blank"` still hits `createWindowOpenHandler` in `electron/window-open-policy.ts`, which **always denies** (GHSA-9f4c-93c8-jc8g). After `allow-popups`, the log becomes `[window-open] denied: https://github.com` rather than a Chromium sandbox warning — the click is still a no-op.

Copy still fails without a hub-scoped `clipboard-sanitized-write` grant: `installMediaPermissions()` only allows media, and `installClipboardShim()` does not apply inside the cross-origin iframe.

The same sandbox is also on the Plugins catalog (`plugins-tab.tsx`) and Bot Mode (`hermes-bots/skills-hub.tsx`), not only `embedded-hub-picker.tsx`.

Wrote up the three layers, logs, and a least-privilege main-process follow-up on #91612. Suggest landing this PR together with that Electron exception (always `action: "deny"`, hub **frame origin** → existing `openExternalUrl`; do not key off `details.referrer` because the hub links are `rel="noreferrer"`).

sylvesterkaczmarek:
Addressed the current-Desktop gaps in signed merge commit eb762bbb. I brought the branch onto current main and completed the fix at all three layers:

- all three Hub/catalog iframes now request clipboard-write plus the popup capabilities;
- clipboard-sanitized-write is granted only for the exact Hermes Hub origin, never clipboard-read;
- setWindowOpenHandler still returns deny for every popup, but a focused frame from the exact Hub origin may hand only http/https/mailto URLs to the existing audited openExternalUrl path;
- opaque/artifact and other origins remain side-effect free.

Added regressions for trusted vs opaque/untrusted opener origins, browser-safe URL schemes, Hub-only clipboard permission, and all three iframe surfaces.

Verification on Node 24: 8/8 focused tests pass, desktop TypeScript typecheck passes, git diff --check passes, and targeted ESLint has no errors (only pre-existing main.ts padding warnings).

unsupportedpastels (commented):
Thanks for expanding this to the Plugins catalog and Bot Mode sibling surfaces; covering all three instances of the same iframe bug looks appropriately scoped.

One non-blocking security suggestion before merge: could the link path use a narrow `postMessage` contract from the embedded site instead of adding an origin-gated side effect to `setWindowOpenHandler` and granting `allow-popups-to-escape-sandbox`? For example, picker mode could intercept the authored links and post `{ type: "hermes-open-link", url }`; the parent would validate `event.origin`, allow only `http:`, `https:`, and `mailto:`, then route through the existing audited opener. That would preserve the CVE boundary's useful invariant that the window-open handler is deny-only and side-effect-free, while keeping popup capability disabled for the iframe.

I realize this requires a coordinated website change/deploy, whereas the current approach is desktop-only. If maintainers prefer avoiding that dependency, the exact-origin checks and scheme filtering here make the current implementation a reasonable fallback. The origin-scoped clipboard work can remain as-is (or be split if a smaller security review is preferred).

sylvesterkaczmarek:
Thanks, this is a useful suggestion. I agree a narrow postMessage contract would be cleaner long term because it would keep the window-open handler fully side-effect-free and let the iframe drop the popup capabilities.

For this PR I would keep the desktop-only fallback so the fix does not depend on a coordinated Hub website deployment: the handler still returns deny for every popup, the side effect is restricted to the exact Hub frame origin, and only http/https/mailto reach the existing audited opener. I would be happy to move to the postMessage model in a follow-up once the website side can land with it.

PR #89129, opened 2026-08-18
fix(models): support OpenRouter preset references

mbac:
## What does this PR do?

Allows native OpenRouter model switches to use account-scoped preset references without being rejected by `/v1/models` validation.

OpenRouter documents two model-field forms:

- `@preset/<slug>` — accepted without probing because account presets never appear in the public model list
- `<model>@preset/<slug>` — validates the base model normally, while preserving the preset suffix through auto-correction

Malformed/non-URL-safe preset slugs are rejected before probing; other providers and custom endpoints keep their existing validation behavior.

## Related Issue

Addresses the OpenRouter preset-validation portion of #31739. This does **not** close that broader issue, which also requests unrelated provider/config work.

Related but not duplicate: #42184 suppresses warnings for declared models on custom providers with `discover_models: false`; this PR fixes native OpenRouter preset references in `validate_requested_model`.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `hermes_cli/models.py`: validate preset slugs as RFC-style URL-safe identifiers; recognize direct OpenRouter presets before live model-list probing; validate combined references by their base model; preserve preset suffixes when correcting model IDs.
- `tests/hermes_cli/test_openrouter_preset_validation.py`: cover valid/invalid slug syntax, direct and combined references, provider isolation, custom endpoints, online/offline auto-correction, and a real config → alias → runtime-provider → `switch_model` path using temporary `HERMES_HOME`.

## How to Test

1. Run `python -m pytest tests/hermes_cli/test_openrouter_preset_validation.py -q -o 'addopts='`.
2. Run `python -m pytest tests/hermes_cli/test_openrouter_preset_validation.py tests/hermes_cli/test_model_validation.py tests/hermes_cli/test_models.py -q -o 'addopts='`.
3. Confirm an alias resolving to `@preset/<slug>` switches successfully with an empty warning message.

## Verification

- Regression sabotage: the new suite fails when the production change is temporarily removed, then passes after restoration.
- Targeted model suites: **75 passed**.
- `git diff --check`: passed.
- Python compilation: passed.
[...]

PR #97951, opened 2026-08-29
fix(models): validate OpenRouter presets structurally on custom endpoints (#97907)

### LINKED ISSUE #97907: [Bug]: OpenRouter preset validation causes models.json preview in Telegram
## Bug

When a preset created in the OpenRouter dashboard is selected through Hermes as `@preset/<slug>`, Hermes validates that identifier against OpenRouter's public `/api/v1/models` catalog.

The preset is a server-side OpenRouter dashboard object, not a Hermes-defined preset or a catalog model ID. Dashboard presets do not appear in `/api/v1/models`, so the switch succeeds but Hermes emits a false “not found in this custom endpoint's model listing” warning containing the raw catalog URL. Telegram fetches that URL for its normal link-preview behavior and presents the large JSON response as a `models.json` document on every preset switch.

## Reproduction

1. Create a preset in the OpenRouter dashboard.
2. Configure an OpenRouter-compatible custom provider in Hermes and add the preset's `@preset/<slug>` identifier to its selectable models.
3. From Telegram, switch to that preset with `/model`.
4. The confirmation contains the catalog warning and Telegram displays `models.json`.

## Expected

Hermes should handle OpenRouter dashboard preset identifiers as server-side presets rather than treating them as ordinary catalog model IDs. The successful switch should not produce a false catalog warning or expose the `/models` response as a Telegram document. Normal Telegram link previews should remain enabled.

Tranquil-Flow:
## What

OpenRouter dashboard presets (`@preset/<slug>`) are account-scoped server-side objects — they never appear in a `/models` listing. On OpenRouter-compatible **custom providers**, `validate_requested_model()`'s custom-endpoint branch probed the catalog, always missed, and soft-accepted with a false "not found in this custom endpoint's model listing" warning embedding the raw catalog URL. Telegram's link preview fetched that URL and rendered the large JSON response as a `models.json` document on every preset switch.

Note the premise: `validate_requested_model()` reclassifies `openrouter` + non-openrouter.ai `base_url` to `custom` (models.py:6508-6509), so the custom branch is exactly the path this issue reports.

## Fix

Handle preset references structurally in the custom branch instead of catalog-probing them:

- **Bare `@preset/<slug>`** — accepted without probing (OpenRouter validates the slug at request time), with a URL-free message. Nothing for a link preview to fetch.
- **`<model>@preset/<slug>`** — the BASE model is validated against the live listing; the preset suffix is preserved verbatim on the persisted/corrected id (same rule as the existing `:nitro` routing-variant handling, which the fuzzy corrector would otherwise strip).
- **Malformed slug** (empty / invalid charset) — rejected with the slug contract (`[A-Za-z0-9._~-]+`), not a misleading catalog warning.
- **Multiple `@preset/` markers** — not a defined form; keeps existing behavior.
- Non-preset identifiers keep the existing custom-branch behavior untouched.

No gateway/Telegram adapter change: the document trigger is removed at the source (no URL in the message), and normal link previews remain enabled per the issue's Expected section.

## Verification

- `tests/hermes_cli/test_openrouter_preset_custom_validation.py` — 13 new tests covering all four layers (bare preset, combined form, malformed slug, non-preset unchanged).
- **Fail-without-fix proven**: on clean upstream/main the suite runs 8 failed / 5 passed (the 5 passing pin unchanged non-preset behavior); with the fix 13/13 pass. Re-verified independently by the reviewer on upstream `0d02f0d1dc`.
- Nearby suites (`test_model_validation.py`, `test_model_switch_custom_providers.py`, `test_custom_provider_model_switch.py`, `test_user_providers_model_switch.py`, `test_openai_codex_model_validation_fallback.py`): 140 passed, 2 failed — both failures reproduce identically on clean upstream/main (pre-existing, unrelated).
- Branch is exactly one commit ahead of upstream/main (`git rev-list --left-right --count` → `0 1`).

## Competitor analysis

Open PR #89129 "fix(models): support OpenRouter preset references" handles `@preset/` only in the `normalized == "openrouter"` branch — correct for native OpenRouter, but the custom-provider scenario of this issue never satisfies that gate (it is reclassified to `custom` upstream of the branch). Scopes are complementary with no gated-code-path overlap; the slug-charset rule mirrors #89129's for cross-branch consistency.

Fixes #97907.

Auto-published by Moonsong via Path B automated pipeline.

### DIFF
--- a/hermes_cli/models.py
+++ b/hermes_cli/models.py
@@ -3359,6 +3359,38 @@ def _openrouter_variant_base(model_id: str) -> Optional[str]:
         return base
     return None
 
+
+# OpenRouter dashboard presets are account-scoped server-side objects
+# (created in the OpenRouter dashboard), referenced as ``@preset/<slug>`` or
+# combined with a base model as ``<model>@preset/<slug>``. They never appear
+# in ``/models`` listings — OpenRouter resolves and validates the slug at
+# request time. (#97907)
+_OPENROUTER_PRESET_MARKER = "@preset/"
+_PRESET_SLUG_RE = re.compile(r"[A-Za-z0-9._~-]+")
+
+
+def _split_openrouter_preset(model_id: str) -> tuple[str, Optional[str]]:
+    """Split an OpenRouter dashboard preset reference into base and suffix.
+
+    Returns ``(base_model, suffix)`` where ``suffix`` is ``"@preset/<slug>"``
+    when exactly one marker is present and the slug is non-empty:
+
+      ``@preset/<slug>``        → ``("", "@preset/<slug>")`` — bare preset
+      ``<model>@preset/<slug>`` → ``("<model>", "@preset/<slug>")``
+      anything else             → ``(model_id, None)`` — not a preset form
+
+    Multiple markers (``a@preset/x@preset/y``) are not a defined OpenRouter
+    form and return ``(model_id, None)`` so callers keep their existing
+    behavior for ordinary identifiers. A single marker with an empty slug
+    (``@preset/``) IS returned as a preset form — the caller's slug-charset
+    check rejects it with the preset contract.
+    """
+    text = model_id or ""
+    if text.count(_OPENROUTER_PRESET_MARKER) != 1:
+        return text, None
+    base, _, slug = text.partition(_OPENROUTER_PRESET_MARKER)
+    return base, f"{_OPENROUTER_PRESET_MARKER}{slug}"
+
 # Subscription/OAuth providers whose catalogs RE-EXPOSE other vendors' models
 # would be listed here (tried only as a last resort for bare short-alias
 # resolution, after every native-vendor catalog, so they never hijack an alias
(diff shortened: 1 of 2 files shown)


Enough1122:

### PR #97951 — fix(models): validate OpenRouter presets structurally on custom endpoints

**Summary:** OpenRouter `@preset/<slug>` identifiers are server-side objects never present in `/models`. Adds `_split_openrouter_preset` and `_PRESET_SLUG_RE` (letters/digits/`._~-`) and structural handling in `validate_requested_model` for `custom` providers: slug charset is validated, bare `@preset/...` is accepted without a probe (no URL in message to trigger Telegram link previews), combined `<model>@preset/...` validates the base against the listing and preserves suffix through auto-correction (mirrors `:nitro` handling). Multiple markers (`a@preset/x@preset/y`) fall back to ordinary catalog validation.

- `_split_openrouter_preset` (`hermes_cli/models.py:35`) counts exactly one `@preset/` marker; empty slug `@preset/` still returns a suffix and is then rejected by the charset check — correct per contract.
- Suggestion generation now uses `_lookup` (base) not original id, so close-match hints are for the base — appropriate.
- Leading/trailing whitespace in `requested` is not stripped before marker count; upstream `validate_requested_model` trims elsewhere, but if a raw `" @preset/foo"` slipped through it would probe incorrectly. Low risk.

**Verdict:** Well-tested fix for #97907 (four layers: bare, combined, malformed, non-preset unchanged). No blocking issues.

Issue #97906, opened 2026-08-29
[Bug]: Mem0 OSS OpenAI defaults misroute requests and send unsupported GPT-5 parameters


GodsBoy:
### Bug Description

Mem0 OSS setup currently defaults its OpenAI LLM to `gpt-5-mini`, but the generated configuration omits `is_reasoning_model: true`. Hermes pins mem0ai 2.0.10, whose model-name heuristic does not classify that exact model as reasoning, so it sends `temperature` and `top_p`. The OpenAI endpoint rejects those parameters for this model.

The same mem0ai release also lets an ambient `OPENROUTER_API_KEY` override an explicitly configured `provider: openai` LLM. Its OpenAI embedder does not use the OpenRouter key, so memory extraction and embedding can silently target different providers.

Both behaviors are present on current main at `5831d8365aea36fea48f9c3e839827d19ccb9a6c`.

### Steps to Reproduce

1. Build the default Mem0 OSS configuration through `build_oss_config(parse_flags(["--mode", "oss"]))`.
2. Observe that `llm.config.model` is `gpt-5-mini` and `llm.config.is_reasoning_model` is absent.
3. With pinned `mem0ai==2.0.10`, construct its LLM base with `OpenAIConfig(model="gpt-5-mini")` and inspect the supported request parameters.
4. Observe `temperature`, `top_p`, and `max_completion_tokens` in the request parameter set.
5. Set distinct sentinel values for `OPENAI_API_KEY` and `OPENROUTER_API_KEY`, then construct mem0ai's `OpenAILLM` with an explicit OpenAI key and base URL while capturing client arguments.
6. Observe that the LLM client receives the OpenRouter key and `https://openrouter.ai/api/v1`, while `OpenAIEmbedding` receives the OpenAI key and configured OpenAI base URL.

The reproduction uses fake clients and makes no provider network request.

### Expected Behavior

- The generated default config marks the exact `gpt-5-mini` model as reasoning so mem0ai omits unsupported sampling parameters.
- A Mem0 OSS LLM explicitly configured with `provider: openai` uses its OpenAI credential and OpenAI-compatible base URL, regardless of an unrelated ambient OpenRouter credential.
- The saved provider identity, embedder config, and process environment remain unchanged.

### Actual Behavior

The reasoning probe reports:

```text
{'reasoning_detected': False, 'request_param_keys': ['max_completion_tokens', 'messages', 'temperature', 'top_p']}
```

The provider probe reports:

```text
[...]

PR #86167, opened 2026-08-14
fix(memory/mem0): resolve qdrant default path from HERMES_HOME at use time

shashb27:
## What does this PR do?

Fixes #85830: the mem0 qdrant default storage path was frozen to `~/.hermes/mem0_qdrant` at **module import time** (`os.path.expanduser` baked into the module-level `VECTOR_PROVIDERS` dict). Because the path derived from `HOME` — not `HERMES_HOME` — the CLI's profile redirection (which sets `HERMES_HOME` before argparse runs) could never reach it. A user running with a redirected `HERMES_HOME` still got their vector store under `$HOME/.hermes/`, defeating per-profile isolation in exactly the default-install case where users expect the redirect to work.

The fix removes the frozen path from the registry and resolves it **lazily at use time** from `get_hermes_home()` (the same canonical helper the rest of the CLI uses). Backward-compatible: explicit `path`/`url` configs and remote-url mode are unchanged.

## Related Issue

Fixes #85830

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✅ Tests (adding or improving test coverage)

## Changes Made

- `plugins/memory/mem0/_oss_providers.py`
  - Dropped the frozen `path` from `VECTOR_PROVIDERS["qdrant"]["default_config"]` → `{}`.
  - Added `default_qdrant_path()` → `str(get_hermes_home() / "mem0_qdrant")`, evaluated at call time.
- `plugins/memory/mem0/_setup.py`
  - `build_oss_config`: no `path` is baked into `mem0.json` unless `--oss-vector-path` is passed; `--oss-vector-url` explicitly drops any local path.
  - `_vector_description`: the picker now shows the resolved HERMES_HOME-relative default instead of the stale "local storage" string.
  - `_run_connectivity_checks`: local-mode check resolves the default lazily when the config carries no explicit path/url.
- `plugins/memory/mem0/_backend.py`
  - `OSSBackend.__init__`: when qdrant is selected and the config has no `path` and no `url`, fills `vs_config["path"]` from `default_qdrant_path()` before the existing `expanduser` call. Covers hand-edited configs that omit the path.

## How to Test

1. `git checkout` this branch, `cd` into the repo.
2. `pytest tests/plugins/memory/test_mem0_providers.py tests/plugins/memory/test_mem0_setup.py tests/plugins/memory/test_mem0_backend.py -q` — all pass.
3. **Regression proof for #85830:**
   ```bash
[...]

Issue #85830, opened 2026-08-14
mem0 qdrant default path is frozen from HOME at import time, so HERMES_HOME redirection cannot reach it


imchul-oss:
## Where

`plugins/memory/mem0/_oss_providers.py`:

```python
VECTOR_PROVIDERS: dict[str, dict[str, Any]] = {
    "qdrant": {
        "label": "Qdrant",
        "default_config": {"path": os.path.expanduser("~/.hermes/mem0_qdrant")},
        ...
```

## Problem

Two things combine:

1. `os.path.expanduser("~/.hermes/mem0_qdrant")` runs at **module import time**, so the
   path is resolved once and baked into the module-level dict.
2. It derives from `HOME`, not from `HERMES_HOME`. The CLI's profile mechanism
   (`hermes_cli/_parser.py` sets `HERMES_HOME` before argparse runs) redirects the rest of
   the runtime, but this default cannot follow it even in principle - a user running with a
   redirected `HERMES_HOME` still gets their vector store under `$HOME/.hermes/`.

The call site (`_backend.py`) re-expands a path coming from config at call time, so the
frozen module default only applies when the config carries no explicit path - which is
exactly the default-install case where a user would expect the profile redirect to work.

## Suggested fix

Resolve the default lazily at the point of use, derived from the same home-resolution
helper the rest of the CLI uses, e.g.:

```python
"default_config": {},  # path filled at use: os.path.join(resolve_hermes_home(), "mem0_qdrant")
```

or make the default a callable evaluated when the provider is instantiated.

## Context

Found while auditing import-time path pinning across the codebase (76 sites derive paths at
import; this one is the only vector-store data path among them, and the only one that is
HOME-based rather than HERMES_HOME-derived, so environment-level redirection fixes that
work for the others cannot cover it). PR #15624 introduced the block. Happy to send a PR if
the lazy-resolution direction is acceptable.

PR #63961, opened 2026-07-13
fix(memory): scope Mem0 Qdrant storage to the active profile

nateEc:
## What does this PR do?

Named Hermes profiles use separate `HERMES_HOME` directories, but Mem0's OSS Qdrant default was materialized as the hard-coded `~/.hermes/mem0_qdrant`. A named profile that did not pass an explicit vector path therefore read and wrote the same local vector store as every other profile, defeating profile isolation and potentially surfacing memories from another profile.

This PR resolves the default Qdrant path from `get_hermes_home()` when OSS configuration is built. Explicit `--oss-vector-path` and `--oss-vector-url` values keep their existing precedence.

Migration note: this intentionally does not move data from the previously shared default store. Its ownership may already be ambiguous or mixed across profiles. Users who deliberately want the old location can keep it with an explicit vector path.

## Related Issue

No dedicated issue. I searched the open PRs for Mem0, Qdrant, and profile-isolation fixes and found no matching change.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Security fix
- [ ] Documentation update
- [ ] Tests (adding or improving test coverage)
- [ ] Refactor (no behavior change)
- [ ] New skill (bundled or hub)

## Changes Made

- Remove the process-global Qdrant path from `plugins/memory/mem0/_oss_providers.py`.
- Derive the implicit local Qdrant path from the active profile in `plugins/memory/mem0/_setup.py`.
- Exercise the real `HERMES_HOME` resolver in a same-process profile-switch regression test while retaining coverage for explicit paths and remote URLs.
- Document the profile-scoped default path and show the old shared path only as an explicit compatibility override.

## How to Test

1. Run `scripts/run_tests.sh tests/plugins/memory/test_mem0_backend.py tests/plugins/memory/test_mem0_providers.py tests/plugins/memory/test_mem0_setup.py tests/plugins/memory/test_mem0_v3.py --file-timeout 180`.
2. Confirm all 123 tests pass, including `test_default_qdrant_path_tracks_active_profile`.
3. Run Ruff on the three changed Python files and `python scripts/check-windows-footguns.py --diff origin/main`.

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
[...]

PR #97953, opened 2026-08-29
fix(mem0): scope OSS history db to the active Hermes profile

### LINKED ISSUE #97923: [Bug]: Mem0 OSS history and migration state is shared across profiles
## Bug Description

Mem0 OSS leaves its SQLite history database and telemetry migration Qdrant store under the native home directory instead of the active Hermes profile. Two Hermes profiles therefore share `history.db` and `migrations_qdrant`, even though the memory provider receives a profile-specific `hermes_home`.

Confirmed on current `main` at `5831d8365aea36fea48f9c3e839827d19ccb9a6c` with generic temporary profiles. The SQLite history can contain raw recent messages, so this is also a profile data-isolation problem.

This was found during the required follow-up triage for #97906. The embedded user-memory Qdrant default has an existing report and fixes in #85830, #63961, and #86167. This issue covers the separate Mem0 SDK history and migration state that those changes do not address.

## Steps to Reproduce

1. Set `HOME=/tmp/native-home`, leave `MEM0_DIR` unset, and select `HERMES_HOME=/tmp/profile-alpha`.
2. Build the pinned `mem0ai==2.0.10` `MemoryConfig` used by OSS mode.
3. Record `history_db_path` and the telemetry migration directory used by `Memory`.
4. Repeat in a clean process with `HERMES_HOME=/tmp/profile-beta` and the same native `HOME`.
5. Compare the paths.

Both profiles resolve to `/tmp/native-home/.mem0/history.db` and `/tmp/native-home/.mem0/migrations_qdrant`.

## Expected Behavior

Every active Hermes profile should use distinct Mem0 history and migration storage derived from its own `HERMES_HOME`. Explicit user paths should continue to take precedence, and existing shared data should not be moved automatically because its profile ownership is ambiguous.

## Actual Behavior

Hermes passes `hermes_home` into the memory provider but does not propagate a profile-local Mem0 data directory or `history_db_path` into the OSS backend. The pinned SDK derives both defaults from `MEM0_DIR` or native `HOME`, so profile changes do not isolate them.

## Affected Component

- Configuration (config.yaml, .env, hermes setup)

N/A (CLI only)

## Debug Report

```shell
Current main: 5831d8365aea36fea48f9c3e839827d19ccb9a6c
Pinned dependency: mem0ai==2.0.10
Profile alpha history: /tmp/native-home/.mem0/history.db
Profile beta history:  /tmp/native-home/.mem0/history.db
Profile alpha migration store: /tmp/native-home/.mem0/migrations_qdrant
Profile beta migration store:  /tmp/native-home/.mem0/migrations_qdrant
```

## Operating System

Ubuntu 24.04

## Python Version

Python 3.12.3

## Hermes Version

`5831d8365aea36fea48f9c3e839827d19ccb9a6c`

```shell
No exception is required. The incorrect shared paths are observable from MemoryConfig and Memory initialization.
```

## Root Cause Analysis (optional)

The pinned SDK computes a module-level Mem0 directory from `MEM0_DIR` or native `HOME` at import time. `MemoryConfig.history_db_path` uses that global, and telemetry initializes `migrations_qdrant` from the same global. Hermes' OSS backend passes vector, LLM, and embedder configuration to Mem0 but no explicit profile-local history path.

Mutating `MEM0_DIR` or SDK module globals around construction is unsafe because one Hermes process can serve multiple profiles. A process-global workaround could race or cross-contaminate sessions.

## Proposed Fix (optional)

Pass an explicit profile-local `history_db_path` into each OSS `Memory` configuration. For `migrations_qdrant`, add or use an SDK-supported per-instance data-directory setting rather than mutating process globals. Test two generic temporary profiles with the real pinned SDK and verify both paths remain distinct.

Document how users can retain an explicit legacy shared path. Do not migrate existing shared data automatically.

liuhao1024:
## What does this PR do?

Scopes the Mem0 OSS `history.db` to the active Hermes profile. mem0ai (pinned 2.0.10) derives `MemoryConfig.history_db_path` from `MEM0_DIR` or the native home **at import time**, and Hermes passed no explicit path into the OSS `Memory` config — so two Hermes profiles sharing a native `HOME` (or one process serving several profiles) silently shared one `history.db`, which contains raw recent messages. That is the profile data-isolation problem from #97923.

The fix threads the `MemoryProvider.initialize()` contract's `hermes_home` (already guaranteed in the kwargs) through `_create_backend()` into `OSSBackend`, and passes an explicit profile-local `history_db_path` (`<hermes_home>/mem0/history.db`) into `Memory.from_config()`. This is a per-instance config key on the SDK's own `MemoryConfig`, so it does not touch `MEM0_DIR` or any SDK module global — safe for multi-profile processes. The parent directory is created up front because mem0ai's `SQLiteManager` does not create parent dirs.

Precedence (all covered by tests):
1. An explicit `history_db_path` in the `oss` config block wins verbatim.
2. An operator-set `MEM0_DIR` env var is respected — nothing is injected and the SDK default (`$MEM0_DIR/history.db`) applies, which is the documented way to retain a legacy shared path.
3. Otherwise the profile-local path is injected.

No existing shared data is moved or migrated — profile ownership of the old shared file is ambiguous, per the issue's expectation.

Scope note: the issue also names `migrations_qdrant`. In mem0ai 2.0.10 that directory is hardcoded inside `Memory.__init__` against the SDK's module-level `mem0_dir` with **no per-instance configuration surface** (`telemetry_config_dict['path'] = os.path.join(mem0_dir, ...)` in `mem0/memory/main.py`), so it cannot be scoped from the host side without mutating process globals — exactly what the issue rules out. It only holds the SDK's anonymous telemetry identity vectors, not user conversation data; the privacy-relevant surface (`history.db` of raw messages) is fully scoped by this PR. An SDK-side per-instance data-dir setting would let Hermes scope it in one line later.

## Related Issue

Fixes #97923

## Type of Change

## Changes Made

- `plugins/memory/mem0/__init__.py` — capture `hermes_home` from `initialize()` kwargs (constructor default `""`), pass it to `OSSBackend` in `_create_backend()`.
- `plugins/memory/mem0/_backend.py` — `OSSBackend.__init__` accepts `hermes_home`; injects `config["history_db_path"] = <hermes_home>/mem0/history.db` (creating the parent dir) unless an explicit `history_db_path` is configured or `MEM0_DIR` is set.
- `tests/plugins/memory/test_mem0_backend.py` — new `TestOSSHistoryDbScoping` (stubbed SDK: injection, no-hermes_home back-compat, `MEM0_DIR` respect, explicit-config precedence) and `TestOSSHistoryDbRealSDK` (real pinned SDK, skipped without the `mem0` extra).
- `tests/plugins/memory/test_mem0_v3.py` — routing sentinel accepts the new `hermes_home` kwarg.

## How to Test

1. `python -m pytest tests/plugins/memory/test_mem0_backend.py tests/plugins/memory/test_mem0_v3.py tests/plugins/memory/test_mem0_setup.py tests/plugins/memory/test_mem0_providers.py -q` → Observed result: 61 passed, 1 skipped (the real-SDK test skips without the `mem0` extra).
2. With the `mem0` extra installed (`pip install "mem0ai==2.0.10"`), run `python -m pytest tests/plugins/memory/test_mem0_backend.py::TestOSSHistoryDbRealSDK -q` → Observed result: 1 passed. This builds two `OSSBackend`s for two temp profiles sharing one native `HOME` and asserts each `Memory` resolved and opened its own `<profile>/mem0/history.db` with no shared `history.db` under the native home.
3. `ruff check` on the touched files → All checks passed.

### Code

- [x] I've tested on my platform: macOS 15 (arm64); the real-SDK integration test also reproduces the issue's Ubuntu scenario path-wise (same Python/SDK code path, `os.path.join` only)

- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — or N/A (no new config key; `history_db_path` in the oss block is honored if an operator sets one, same as any `MemoryConfig` passthrough)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — paths are built with `os.path.join` from the profile root; no platform-specific behavior added

### DIFF
--- a/plugins/memory/mem0/__init__.py
+++ b/plugins/memory/mem0/__init__.py
@@ -203,6 +203,9 @@ class Mem0MemoryProvider(MemoryProvider):
         self._mode = "platform"
         self._api_key = ""
         self._host = ""
+        # Active HERMES_HOME from initialize() kwargs; scopes OSS-local
+        # storage (mem0 history db) to the running profile (#97923).
+        self._hermes_home = ""
         self._user_id = _DEFAULT_USER_ID
         self._agent_id = "hermes"
         self._rerank_default = False
@@ -280,7 +283,7 @@ class Mem0MemoryProvider(MemoryProvider):
         try:
             if self._mode == "oss":
                 from ._backend import OSSBackend
-                return OSSBackend(self._config.get("oss", {}))
+                return OSSBackend(self._config.get("oss", {}), hermes_home=self._hermes_home)
             if self._host:
                 from ._backend import SelfHostedBackend
                 return SelfHostedBackend(self._api_key, self._host)
@@ -335,6 +338,10 @@ class Mem0MemoryProvider(MemoryProvider):
             )
 
     def initialize(self, session_id: str, **kwargs) -> None:
+        # Profile-scoped storage root: always present per the MemoryProvider
+        # contract. Captured before anything builds a backend so OSS-local
+        # state lands under the active profile, not the native home (#97923).
+        self._hermes_home = kwargs.get("hermes_home") or ""
         self._config = _load_config()
         self._mode = self._config.get("mode", "platform")
         self._api_key = self._config.get("api_key", "")
--- a/plugins/memory/mem0/_backend.py
+++ b/plugins/memory/mem0/_backend.py
@@ -156,7 +156,7 @@ class SelfHostedBackend(Mem0Backend):
 class OSSBackend(Mem0Backend):
     """Wraps mem0.Memory for self-hosted (OSS) mode."""
 
-    def __init__(self, oss_config: dict):
+    def __init__(self, oss_config: dict, hermes_home: str | None = None):
         import os
         from mem0 import Memory
 
@@ -203,6 +203,20 @@ class OSSBackend(Mem0Backend):
             "embedder": _provider_block("embedder"),
             "version": "v1.1",
         }
+        # mem0ai derives history_db_path from MEM0_DIR or the native home at
+        # import time, so every profile in the process would share one
+        # history.db of raw recent messages (#97923). Scope it to the active
+        # Hermes profile unless the operator pinned an explicit path — via the
+        # oss config block or MEM0_DIR, which the SDK honors as its own
+        # default. mem0ai's SQLiteManager does not create parent directories,
+        # so the profile-local one must exist before Memory opens the db.
+        explicit_history_db = oss_config.get("history_db_path")
+        if explicit_history_db:
+            config["history_db_path"] = explicit_history_db
+        elif hermes_home and not os.environ.get("MEM0_DIR"):
+            history_dir = os.path.join(hermes_home, "mem0")
+            os.makedirs(history_dir, exist_ok=True)
+            config["history_db_path"] = os.path.join(history_dir, "history.db")
         self._memory = Memory.from_config(config)
 
     @staticmethod
--- a/tests/plugins/memory/test_mem0_backend.py
+++ b/tests/plugins/memory/test_mem0_backend.py
@@ -153,6 +153,100 @@ class TestOSSBackend:
         assert raw == before
 
 
+def _oss_test_config(tmp_path):
+    """Minimal OSS config: local qdrant under tmp, offline-capable providers."""
+    return {
+        "llm": {"provider": "openai", "config": {"model": "gpt-5-mini", "api_key": "sk-test"}},
+        "embedder": {"provider": "openai", "config": {"model": "text-embedding-3-small", "api_key": "sk-test"}},
+        "vector_store": {"provider": "qdrant", "config": {"path": str(tmp_path / "vs")}},
+    }
+
+
+class TestOSSHistoryDbScoping:
+    """#97923: OSS history.db must land under the active Hermes profile, with
+    operator-pinned paths (oss config key or MEM0_DIR) left to the SDK."""
+
+    def _stub_mem0(self, monkeypatch):
+        import sys
+        import types
+
+        captured = {}
+
+        class Memory:
+            @staticmethod
+            def from_config(config):
+                captured.update(config)
+                return FakeOSSMemory()
+
+        stub_mem0 = types.ModuleType("mem0")
+        stub_mem0.Memory = Memory  # type: ignore[attr-defined]
+        monkeypatch.setitem(sys.modules, "mem0", stub_mem0)
+        return captured
+
+    def test_history_db_scoped_to_hermes_home(self, monkeypatch, tmp_path):
+        captured = self._stub_mem0(monkeypatch)
+        monkeypatch.delenv("MEM0_DIR", raising=False)
+        OSSBackend(_oss_test_config(tmp_path), hermes_home=str(tmp_path / "alpha"))
+        assert captured["history_db_path"] == str(tmp_path / "alpha" / "mem0" / "history.db")
+        # mem0ai's SQLiteManager does not create parent dirs; the fix must.
+        assert (tmp_path / "alpha" / "mem0").is_dir()
+
+    def test_history_db_not_injected_without_hermes_home(self, monkeypatch, tmp_path):
+        captured = self._stub_mem0(monkeypatch)
+        monkeypatch.delenv("MEM0_DIR", raising=False)
+        OSSBackend(_oss_test_config(tmp_path))
+        assert "history_db_path" not in captured
+
+    def test_mem0_dir_env_keeps_sdk_default(self, monkeypatch, tmp_path):
+        captured = self._stub_mem0(monkeypatch)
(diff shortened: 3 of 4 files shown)


andrexibiza (commented):
Reviewed exact head `639b301bad797af1eef2476e79ee4170c4839130` against live `main@b4b7727ea07681b40402de411ddd000bb3c439fc`, the full diff, the pinned mem0ai 2.0.10 implementation, #97923’s acceptance contract, the deleted-profile tombstone invariant, and adjacent Mem0 profile-isolation work.

The core history-db repair is pointed in the right direction. Passing the already-qualified `hermes_home` into `OSSBackend` and using Mem0’s per-instance `history_db_path` avoids the process-global `MEM0_DIR` mutation trap, and the real-SDK two-profile test is valuable. This is complementary to #63961 (`nateEc`) / #86167 (`shashb27`) and #85830 (`imchul-oss`): those own the user-memory Qdrant default; this PR owns Mem0’s separate raw-message history store. Please preserve those lineages rather than treating them as duplicates.

I found two landing blockers.

## P1 — the new mkdir bypasses the repository’s deleted-profile tombstone gate

`OSSBackend.__init__` now does:

```python
history_dir = os.path.join(hermes_home, "mem0")
os.makedirs(history_dir, exist_ok=True)
```

That is not a neutral mkdir under a profile root. Current main deliberately routes profile-owned directory creation through `hermes_constants.mkdir_under_hermes_home()`, which calls `assert_named_profile_home_live()` before creating anything. The contract is explicit: **deleted named profiles must stay gone until explicitly recreated**. `tests/hermes_cli/test_deleted_profile_tombstone.py` exists specifically because stale live work used to recreate `profiles/<name>/...` after `hermes profile delete`, producing zombie profile shells.

This PR reintroduces that other side of the shape for Mem0. Given a tombstoned `hermes_home=/.../profiles/worker`, constructing the OSS backend can recreate `/.../profiles/worker/mem0` directly, bypassing the tombstone check. That is especially unfortunate in a change whose purpose is to make storage obey qualified profile identity.

Required repair: use the canonical guarded mkdir helper (or an equivalent call to `assert_named_profile_home_live()` before creation), and add a regression that tombstones a named profile, attempts Mem0 OSS initialization against that exact `hermes_home`, gets the expected refusal, and proves the profile directory remains absent. Keep the unrelated custom-home control so `/srv/profiles/buildcache`-style paths still work; the current tombstone tests already define that distinction.

## P1 — `Fixes #97923` closes a two-surface issue while one surface is intentionally still shared

#97923’s expected behavior is explicit: every active Hermes profile should get distinct Mem0 **history and migration storage**. The issue reproduces both the raw-message `history.db` and `migrations_qdrant` under the native Mem0 home.

This PR correctly fixes the first half, but its own scope note says the second half remains unresolved because mem0ai 2.0.10 hardcodes the telemetry migration path from module-level `mem0_dir`. I verified that exact dependency source: `Memory.__init__` still sets `collection_name="mem0migrations"`, derives `migrations_qdrant` / `migrations_faiss` from `mem0_dir`, calls `os.makedirs(...)`, and exposes no per-instance data-dir override for that path.

So the implementation scope can stay narrow, but the closure semantics cannot say `Fixes #97923`. Either make this `Part of #97923` and leave the issue open for the SDK-bound migration-store half, or split the remaining acceptance criterion into an explicitly linked follow-up issue before using an auto-close token. The fact that the remaining store is telemetry/migration state rather than raw conversation history changes severity, not whether the issue’s stated contract is complete.

## Interlocks / merge order

- #63961 (`nateEc`) and #86167 (`shashb27`) are adjacent/complementary Qdrant profile-scoping implementations; they should not be overwritten or credited to this change.
- #85830 (`imchul-oss`) is the underlying import-time Qdrant-default report; #97923 was correctly opened as a separate Mem0 SDK-local history/migration class.
- #97906 is a different Mem0 OSS boundary (provider routing / GPT-5 parameter semantics). It shares `_backend.py` as an implementation surface, so whichever lands second needs a fresh current-main composition check even though the defect classes are distinct.
- The reviewed head is one submitted commit, currently 16 commits behind live main. I found no main-only overlap with this PR’s four files in the interval I inspected, but it still needs a current-main refresh before landing because main moved after the branch point.

## Verification

At review time, exact-head Docker `33258337248` and Nix `33258337226` are green. Required CI `33258337508` is still in progress, so this object does not yet have a complete hosted landing receipt. The PR’s focused local/real-SDK results are useful development evidence, but they do not replace the exact-head aggregate.

The history-db isolation mechanism itself is good work: per-instance ownership is exactly the right way to avoid cross-profile process globals. Once the mkdir obeys the tombstone authority boundary, the issue closure is made truthful, and the exact refreshed object is green, this becomes a clean narrow repair. 🚀

PR #97955, opened 2026-08-29
packaging: exempt every exact pin from exclude-newer (release-day update brick)

walker83:
## Problem

Every release exact-pins at least one dependency to a version published days before the release. With the relative `exclude-newer = "14 days"` cutoff, those fresh pins are invisible to the resolver for ~two weeks after release, so any install whose venv predates the release bricks on a plain `hermes update`:

```
× No solution found when resolving dependencies:
╰─▶ Because there is no version of snowballstemmer==3.1.1 and
    hermes-agent==0.20.6 depends on snowballstemmer==3.1.1, ...
hint: `snowballstemmer` was filtered by `exclude-newer` to only include
packages uploaded before 2026-08-15T12:30:31Z.
```

**Field report, 2026-08-29** — updating three long-running installs from v0.20.0 to v0.20.6 (one Termux/Android, two Linux servers), uv hard-failed in sequence on `snowballstemmer==3.1.1`, `psutil==7.2.2`, `certifi==2026.5.20` ("has no publish time"), then extras pins `fastapi`, `uvicorn`, `cffi`/`pycparser`, `python-telegram-bot`. On the Termux host, `cryptography`'s isolated sdist build also bricked under the same cutoff via its `build-system.requires` (`maturin>=1.9,<2`, `setuptools-rust>=1.11.0`). A second install that had been continuously updated since before the release (venv already holding the pinned versions) sailed through — resolution only breaks when the pinned version must actually be downloaded.

This is the same shape as the existing `setuptools`/`pillow`/`mcp`/`firecrawl-anydoc` exemptions: for an exact pin, the cutoff adds zero float protection (the version cannot move without a reviewed pin bump) and can only brick.

## Change

- Extend `[tool.uv].exclude-newer-package` to cover **every exact-pinned package** in `[project].dependencies` + `optional-dependencies` (13 → 97 entries; table moved to one-key-per-line for reviewability — semantically identical TOML).
- Add `maturin` and `setuptools-rust`: `build-system.requires` of the already-exempted `cryptography` pin, needed on wheel-less platforms (Termux/Android) that must build it from sdist.
- Add `tests/test_packaging_metadata.py::test_exact_pinned_deps_exempt_from_exclude_newer`, mirroring the existing build-system test: adding a `name==version` pin without a matching whitelist entry now fails CI, so this stops being per-release whack-a-mole.

## Notes

- Ranged requirements are intentionally **not** exempted — the cutoff keeps its float protection exactly where a fallback candidate exists.
- `uv sync`/`uv pip` behavior for non-pinned deps is unchanged; only exact pins bypass the cutoff.

## Testing

- `pytest tests/test_packaging_metadata.py` — 8 passed (includes the new test; verified it fails against the un-extended whitelist).
- `ruff check` clean on the touched files.
- Applied locally to the v0.20.6 checkout on all three machines above; `hermes update` then completed end-to-end (deps resolved, gateways restarted, messaging platforms reconnected).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -433,7 +433,124 @@ exclude-newer = "14 days"
 # firecrawl-anydoc: same exact-pin shape (==0.2.4, hosted-OCR wiring PR).
 # The pin bump WAS the review; exclude-newer adds zero float protection to
 # an exact pin and would only delay the reviewed version 14 days.
-exclude-newer-package = { vercel = false, nemo-relay = false, huggingface_hub = false, h2 = false, aiohttp = false, cryptography = false, defusedxml = false, python-olm = false, unpaddedbase64 = false, setuptools = false, pillow = false, mcp = false, firecrawl-anydoc = false }
+#
+# Every other exact-pinned package below: the release-day brick shape
+# (observed 2026-08-29 updating three long-running installs v0.20.0 ->
+# v0.20.6 — one Termux, two Linux servers). Each release exact-pins at least
+# one dependency to a version published days before the release (v0.20.6
+# pinned snowballstemmer==3.1.1 and psutil==7.2.2 in core, plus fastapi /
+# uvicorn / python-telegram-bot pins across extras). For two weeks after
+# release the cutoff filters those versions out, so any venv that predates
+# the release bricks on `hermes update` ("no version of
+# snowballstemmer==3.1.1") until the window passes. Same zero-float-
+# protection logic as setuptools/pillow/mcp above; enforced for every exact
+# pin by
+# tests/test_packaging_metadata.py::test_exact_pinned_deps_exempt_from_exclude_newer.
+#
+# maturin, setuptools-rust: build-system.requires of the cryptography pin
+# above. cryptography itself is exempted, but on wheel-less platforms
+# (Termux/Android) it must build from sdist, and the isolated build
+# environment resolves under the same cutoff ("Failed to resolve
+# requirements from build-system.requires ... maturin>=1.9,<2").
+[tool.uv.exclude-newer-package]
+agent-client-protocol = false
+ai-edge-litert = false
+aiohttp = false
+aiohttp-socks = false
+aiosqlite = false
+alibabacloud-dingtalk = false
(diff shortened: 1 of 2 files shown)


Enough1122:

### PR #97955 — packaging: exempt every exact pin from exclude-newer (release-day update brick)

**Summary:** Replaces the 12-entry inline `exclude-newer-package` with a 70+ entry `[tool.uv.exclude-newer-package]` whitelist (`agent-client-protocol = false`, …, `youtube-transcript-api = false`). Covers every exact-pinned dep in `[project].dependencies`/`optional-dependencies` plus build-system sdist deps (`maturin`, `setuptools-rust`) needed on wheel-less platforms (Termux). Adds `test_exact_pinned_deps_exempt_from_exclude_newer` which loads `pyproject.toml` and asserts no exact pin is missing from the whitelist while a relative `exclude-newer = "14 days"` cutoff is configured.

- Whitelist is manually maintained but now regression-guarded by the new test — missing entries will fail CI. Consider generating it from `_pins_from_specs` to avoid drift, but current approach is explicit.
- `uv` canonicalization (`_canonical`) handles name normalization (`python-olm` vs `python_olm`) so the check is robust.
- Expanding to every exact pin adds zero safety loss: the pin bump was the review; `exclude-newer` with an exact pin can only brick (observed 2026-08-29 `snowballstemmer==3.1.1` / `psutil==7.2.2` across 3 long-lived installs).

**Verdict:** Correct packaging fix for the observed release-day brick class. No functional issues.

teknium1:
Superseded by salvage PR #99493, which carries your commits (the 98-entry exact-pin `exclude-newer-package` table + the guard test) onto current main **with your authorship preserved** via cherry-pick, plus the `contributors/emails` mapping for your commit email.

Your field report of the release-day brick (snowballstemmer/psutil/certifi chain across three installs) was the definitive evidence for the class fix. #99493 additionally regenerates `uv.lock` — the exclusion-table change makes uv refuse `lock --check`/`--locked` syncs until the lock's options block matches (verified metadata-only regen, zero version/hash changes). Thanks @walker83!

PR #86049, opened 2026-08-14
fix(gateway): route synthetic auto-TTS through platform adapters

0xbWy:
## What does this PR do?

Delayed/background completions are re-injected as synthetic `MessageEvent`s without a raw Discord guild object. `_send_voice_reply()` therefore could not use its direct guild playback path and called `send_voice()` directly, bypassing Discord's existing text-channel → connected-voice-channel binding.

This change delegates the no-direct-guild case through the platform adapter's existing `play_tts(chat_id, ...)` contract. Discord can then resolve its bound voice channel; platforms or Discord chats without an active binding retain their existing attachment fallback. Direct raw-guild playback remains first.

## Related Issue

No existing issue found after searching open and closed issues/PRs for synthetic, delayed, and background completion TTS/voice delivery. This fixes a live-reproduced gateway bug reported through Discord.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `gateway/run.py`: use `adapter.play_tts()` when direct guild voice playback is unavailable, preserving `send_voice()` compatibility for legacy adapters.
- `tests/gateway/test_async_completion_voice_routing.py`: cover synthetic completion routing, real Discord bound-VC resolution, raw-guild priority, adapter attachment fallback, and legacy compatibility.
- `tests/gateway/test_auto_voice_reply_format.py`: assert auto-TTS uses the adapter playback contract.
- `tests/gateway/test_voice_command.py`: update gateway voice expectations and stale Discord routing descriptions.

## How to Test

1. Enable Discord auto-TTS and bind a text channel to a connected voice channel.
2. Trigger a background/sub-agent completion that returns after the initiating assistant turn has ended.
3. Confirm the delayed completion plays directly in the bound VC rather than appearing only as text or a `voice-message.ogg` attachment.
4. Disconnect the VC or remove the binding and confirm normal attachment fallback remains available.
5. Run `scripts/run_tests.sh` for CI parity.

Automated verification:

- Canonical `scripts/run_tests.sh` full-suite run completed in a clean external `[all,dev]` environment.
- Full-suite result: 22 unrelated failing files / 80 tests.
[...]

PR #97956, opened 2026-08-29
fix(discord): background-completion finals speak in the bound live voice session

jarvis-marcadaxstudios:
## Problem

When the bot is connected to a Discord voice channel, a final response produced by a **background completion** is delivered as a text-channel audio attachment instead of being spoken in the voice channel the bot is already sitting in — even though a reply to a normal message in that same chat, seconds earlier, is spoken correctly.

From a user's point of view: you are in a voice call with the bot, you ask it to run something long, and when the work finishes the answer silently appears as a file in the text channel rather than being spoken.

This affects every final that is synthesised rather than triggered by a real inbound message:

- background process completions (`notify_on_complete`)
- watch-pattern notification injections
- async delegation completions

## Root cause

`_send_voice_reply` resolves the target guild through `_get_guild_id(event)`, which reads `event.raw_message`:

```python
def _get_guild_id(event: MessageEvent) -> Optional[int]:
    """Extract Discord guild_id from the raw message object."""
    raw = getattr(event, "raw_message", None)
    if raw is None:
        return None
    ...
```

A synthesised final has **no raw platform message** — `_inject_watch_notification` and the async-delegation path construct the event without one, because no Discord message caused it. So `_get_guild_id` returns `None`, the voice path is skipped, and delivery falls back to an attachment.

The information needed was already available: `DiscordAdapter` maintains `_voice_text_channels` (guild → bound text channel), which `DiscordAdapter.play_tts` already uses for the voice-origin path.

## Approach

Add `_resolve_voice_playback_guild(event, adapter)`, which keeps `_get_guild_id` as the primary source and falls back to the adapter's existing voice↔text binding.

The fallback is deliberately narrow. It resolves a guild only when **all** hold:

- the event is on the Discord platform
- a binding exists whose bound text channel **is** this chat
- the bot is actually connected to voice for that guild

An unbound chat, a stale binding, or a dropped connection all keep the existing attachment behaviour. No configuration is introduced and no default changes: a chat with no live voice session behaves exactly as before.

## Tests

17 tests in `tests/gateway/test_discord_background_completion_voice.py`, covering the fix and the negative space around it: voice-origin finals, typed finals in a bound session, background-completion finals, resolver chat matching, unbound chats, no binding at all, disconnected voice clients, stale bindings, non-Discord platforms, replayed deliveries (played once, never doubled), and playback failure not also emitting an attachment.

Verified to actually bite: with `gateway/run.py` reverted to `main` and the new tests kept, **8 of the 17 fail**; with the fix applied all 17 pass.

Also run on this branch:

- `pytest tests/gateway -k "discord or voice"` → 501 passed, 4 failed
- Those same 4 (`test_discord_send.py` ×3, `test_send_multiple_images.py` ×1) fail identically on pristine `main` with this branch's changes absent, and pass when those two files are run alone. They are a pre-existing cross-file isolation issue, unrelated to this change.
- `ruff check` on both changed files → clean

## Compatibility and scope

- No new configuration keys, no behaviour change for anyone not in a voice channel.
- Only `_send_voice_reply`'s guild resolution changes; the existing TTS, mixer and attachment paths are untouched.
- 63 non-test lines changed in one file.

## Limitations

- Only the Discord adapter is addressed. Other platforms have no equivalent voice↔text binding, so the fallback returns `None` for them and behaviour is unchanged.
- Resolution is by exact chat match. A final routed to a *thread* under a voice-bound parent channel is not covered, and keeps the existing attachment behaviour.

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -22896,6 +22896,64 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
             return raw.guild.id
         return None
 
+    def _resolve_voice_playback_guild(self, event: MessageEvent, adapter) -> Optional[int]:
+        """Guild whose live voice session should receive this turn's TTS.
+
+        ``_get_guild_id`` reads the raw platform message, which only exists for
+        real inbound Discord messages and slash interactions. Finals produced by
+        a background process completion, a watch-pattern injection or an async
+        delegation are synthesised WITHOUT a ``raw_message`` (see
+        ``_inject_watch_notification``), so the guild resolved to None and
+        ``_send_voice_reply`` delivered the reply as a text-channel audio
+        attachment even while the bot was speaking in that chat's bound voice
+        channel seconds earlier.
+
+        Fall back to the adapter's voice<->text binding — the very mapping
+        ``DiscordAdapter.play_tts`` uses for the voice-origin path, and the
+        voice-ack resolver in the turn runner uses for mid-turn speech.
+        Deliberately narrow: only a guild whose bound text channel IS this
+        chat, and only while the bot is actually connected there. An unbound
+        chat, a stale binding, or a dropped connection keeps the existing
+        attachment delivery policy.
+        """
+        guild_id = self._get_guild_id(event)
+        if guild_id is not None:
+            return guild_id
+        source = getattr(event, "source", None)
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: #86049 targets the same synthetic Discord completion TTS fallback through adapter routing. This PR resolves the existing voice-to-text binding directly; these are competing mechanisms for maintainers to compare.

PR #97958, opened 2026-08-29
feat(loop): add --start-now to run the first iteration immediately

### LINKED ISSUE #97954: [Feature]: /loop --start-now flag to run the first iteration immediately, then recurrently
### Problem or Use Case

`/loop [interval] <prompt>` schedules the first wakeup one full interval after the command runs — `LoopManager.set()` sets `next_due_at = now + interval`. When I type `/loop 1h check the deploy status`, I just told Hermes what to check; the natural expectation is to run it once **now**, then again every hour. Waiting the full interval before any output feels like the command was ignored.

Claude Code parity is acknowledged — its `/loop` also waits the interval first — but parity isn't necessarily the desired behavior for every user.

### Proposed Solution

Add an opt-in flag, `--start-now`, to `/loop`:

- **Default (no flag):** unchanged — first wakeup after one interval (Claude Code parity).
- **`--start-now`:** first wakeup fires immediately (on the next idle poll / gateway watcher scan), then continues on the interval.

```
/loop 1h check the deploy status                        # first run in 1h (unchanged)
/loop 1h --start-now check the deploy status            # first run now, then hourly
```

Works for both cadence modes: fixed interval and self-paced.

**Implementation sketch** (`hermes_cli/loops.py`):
- `parse_loop_args()`: parse and strip `--start-now` (alongside `--times`/`--until`); add `"start_now": bool` to the return dict.
- `LoopState`: new persisted field `start_now: bool = False`.
- `LoopManager.set()`: when `start_now`, set `next_due_at = now` instead of `now + interval` (self-paced: also `now`).
- `resume()` unchanged — already re-arms relative to now.
- Pass-through in CLI (`cli.py`), gateway (`gateway/run.py`), TUI.
- Update `/loop` help text + `website/docs/user-guide/features/loops.md` ("First wakeup in 5m" → note the flag).
- Tests in `tests/hermes_cli/test_loops.py` + gateway loop test.

The `/proactive` alias shares the same code path, so the flag works there too.

### Alternatives Considered

1. **Change the default to run-now** — rejected: breaks Claude Code parity and surprises existing users; a flag keeps it backwards-compatible.
2. **`--now` boolean flag** — rejected for `--start-now` per user preference; explicit about intent.
3. **Run the task once manually before `/loop`** — doesn't scale to unattended gateway use.

### Feature Type

CLI improvement

### Scope

Small (single file, < 50 lines)

### Contribution

### Debug Report (optional)

```shell

```

oxngon:
## Summary

Adds an opt-in `--start-now` flag to `/loop` so the first iteration fires immediately, then recurs on the normal cadence. Default behavior is unchanged (first wakeup after one interval — Claude Code parity).

```
/loop 1h check the deploy status            # first run in 1h (unchanged)
/loop 1h --start-now check the deploy       # first run NOW, then hourly
```

## Motivation

`/loop [interval] <prompt>` schedules the first wakeup one full interval after the command runs. When you've just told Hermes what to check, waiting the whole interval before any output feels like the command was ignored. This adds an explicit opt-in to run immediately, keeping the parity default for everyone else.

Closes #97954

## Changes

- `parse_loop_args()`: parse and strip `--start-now` (works leading or trailing, plays nice with `--times`/`--until`); add `start_now` to the return dict.
- `LoopState`: new persisted `start_now` field (default `False`; survives serialization round-trip and old rows missing the field).
- `LoopManager.set()`: set `next_due_at = now` when `start_now`, for both fixed-interval and self-paced modes.
- `dispatch_loop_command()`: wire `start_now` through, update `/loop` he

### DIFF
--- a/hermes_cli/loops.py
+++ b/hermes_cli/loops.py
@@ -149,11 +149,14 @@ def parse_loop_args(text: str) -> Dict[str, Any]:
         /loop keep fixing the failing test until the suite passes
         /loop 2m poll CI --times 30
         /loop 5m watch the queue --until queue depth reaches zero
+        /loop 1h --start-now check the deploy status
 
     Returns ``{"interval_seconds": int|None, "prompt": str, "times": int,
-    "until": str, "error": str|None}``. ``interval_seconds`` None means
-    self-paced. ``error`` is set for unusable input (empty prompt,
-    interval-only, bad --times).
+    "until": str, "start_now": bool, "error": str|None}``.
+    ``interval_seconds`` None means self-paced. ``start_now`` is True when
+    the user passed ``--start-now`` (first wakeup fires immediately).
+    ``error`` is set for unusable input (empty prompt, interval-only,
+    bad --times).
     """
     raw = (text or "").strip()
     result: Dict[str, Any] = {
@@ -161,6 +164,7 @@ def parse_loop_args(text: str) -> Dict[str, Any]:
         "prompt": "",
         "times": 0,
         "until": "",
+        "start_now": False,
         "error": None,
     }
     if not raw:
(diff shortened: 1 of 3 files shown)


teknium1:
Merged via PR #97980 — your commit was cherry-picked onto current main with your authorship preserved in git log. Per maintainer direction we flipped the behavior from an opt-in `--start-now` flag to the unconditional default: `/loop`'s first iteration now always fires immediately, then recurs on the cadence (the flag was never released, so it was dropped rather than deprecated). Thanks for the contribution — the parser/state/tests/docs groundwork all carried over.

PR #97491, opened 2026-08-28
fix(cron): preserve schedule timezone across processes

tinyloop-farid-codex:
## Summary

This refreshes #88581 onto current `main` while preserving @JoaoMarcos44's two original commits and authorship.

Cron expressions describe local wall-clock intent, but persisted `next_run_at` values did not record which timezone that wall clock belonged to. A long-running gateway and a fresh CLI/web worker could therefore interpret the same schedule differently after a timezone configuration change. The fix stamps cron schedules with their IANA timezone, evaluates future occurrences in that zone, and refreshes the process timezone cache safely.

Current `main` also added a stale-expression guard after #88581 was opened. The integration commit makes that guard validate stamped timestamps in the schedule timezone; otherwise every non-UTC stamped run is incorrectly treated as stale and deferred instead of firing.

The refreshed branch also preserves pending `run now` intent while a timezone rebase is waiting. The rebase now waits until the manual marker and any one-run prompt have been consumed instead of silently replacing the operator request.

Fixes #88220.

## Verification

- `scripts/run_tests.sh tests/cron/test_cron_schedule_timezone_88220.py tests/cron/test_jobs.py tests/cron/test_due_stale_cron_edit.py tests/cron/test_cron_relay_run_forward.py -q` — 161 passed
- `scripts/run_tests.sh tests/test_hermes_time.py -q` — 5 passed
- `.venv/bin/ruff check cron/jobs.py hermes_time.py tests/cron/test_cron_schedule_timezone_88220.py tests/test_hermes_time.py` — clean

Agent model / effort: GPT-5 / runtime-managed

— posted by Codex

PR #92489, opened 2026-08-22
fix(timezone): isolate cache by active profile

qdivan:
## Summary
- Key `hermes_time`’s cached timezone by the active source: the explicit environment override or the profile-specific `config.yaml` path.
- Prevent a multiplexed process that switches `HERMES_HOME` from returning the previous profile’s timezone.
- Add a regression test that resolves Tokyo under one temporary home, switches to a second home, and requires New York without a manual cache reset.

## Validation
- `scripts/run_tests.sh tests/test_timezone.py -q` (12 passed)
- `scripts/run_tests.sh tests/hermes_cli/test_managed_scope_loaders.py -q` (2 passed)
- `git diff --check`
- `python3 -m py_compile hermes_time.py tests/test_timezone.py`

Related to the profile-isolation cache boundary; no existing issue covers this exact behavior.

PR #97961, opened 2026-08-29
fix(cron): re-resolve hermes_time per profile in multiplex ticker

### LINKED ISSUE #97905: Cron: `next_run_at` computed under a foreign process's timezone persists wrong-offset fire times into profile jobs.json
## Component
`cron/jobs.py` — `next_run_at` persistence + multiplex/desktop tick interaction with `hermes_time`.

## Environment
- Hermes Agent v0.20.0 (2026.8.3), Linux, 8 profiles, per-profile gateways as systemd user services + Hermes desktop app
- Profile `north-caribbean` with `timezone: America/New_York` in its config.yaml
- Built-in cron provider

## Summary

A profile's cron jobs had their `next_run_at` timestamps persisted with **UTC offsets** despite the profile being configured `America/New_York`, causing every job to fire **4 hours early** (10:00 ET instead of 14:00 ET for `0 14 * * *`).

Root cause chain:

1. When the Hermes desktop app runs, its multiplex cron ticker (`_start_desktop_cron_ticker` in `hermes_cli/web_server.py`) ticks **every** profile's cron store. The tick loop scopes agent execution per-profile via `set_hermes_home_override()`, but the **desktop dashboard backend process itself was started under a different profile context** (its own `HERMES_HOME`) whose `hermes_time` resolution — and in some paths plain `datetime.now()` — returns the desktop backend's timezone (or UTC on a server with `Etc/UTC` system time), not the ticked profile's configured timezone.

2. `compute_next_run()` / `mark_job_run()` calls made during that foreign tick therefore persist `next_run_at` as a UTC-offset instant into the **ticked profile's** `jobs.json`. Evidence: all jobs created/re-armed before the desktop app started carry `+00:00`; a job created after the desktop app closed (by the profile's own gateway) correctly carries `-04:00`.

3. On subsequent ticks — even by the correct profile gateway with the right timezone — the stored UTC instant is compared literally against Eastern `now`. `0 14 * * *` stored as `14:00+00:00` becomes due at **10:00 ET** and fires four hours early.

## The migration guard almost saves it, but has a hole

The offset-mismatch repair in `_get_due_jobs_locked()` (`cron/jobs.py:3616-3638`, from #28934) recomputes `next_run_at` when the stored offset differs from now's — but only when **all four** conditions hold, including `_stored_wall_clock_is_future(raw_next_run_dt, now)`. That wall-clock comparison compares the stored UTC wall clock ("14:00") against the current Eastern wall clock ("13:45") — a mismatched-offset pair — so during the 10:00–14:00 ET window the guard fires correctly. But **before 10:00 ET** the stored value is not yet due at all (`next_run_dt > now` compares instants: 14:00Z = 10:00 ET), so no repair happens and the job simply fires early at 10:00 ET when it first becomes "due" — the guard only helps for occurrences *after* the first wrong fire. Each foreign-tick re-arm re-introduces the problem.

Also worth noting: the guard's wall-clock comparison treats a UTC-stored time as if its wall clock were local — `14:00+00:00` read against an `America/New_York` clock as "2 PM is still future" — which is only coincidentally the right answer. A profile at a negative offset where the stored wall clock is behind local would take the opposite (also wrong) path.

## Repro

1. Configure profile A with `timezone: America/New_York` on a host whose system tz is UTC, running A's gateway.
2. Start the Hermes desktop app while its active profile is B (any other profile, or system-UTC default).
3. Create or let fire a cron job in A with a wall-clock schedule (e.g. `0 14 * * *`).
4. Observe the job's `next_run_at` in A's `cron/jobs.json` carries `+00:00` and the job fires at 10:00 ET.

Deterministic variant: stop A's gateway so only the desktop ticker can fire A's jobs.

## Expected behavior

`next_run_at` for a profile's job must always be computed in the **job-owning profile's** configured timezone, regardless of which process's ticker computes it. Two candidate fixes:

1. The multiplex tick already scopes the agent home via `set_hermes_home_override()` / `use_cron_store()` — the timezone resolution used by `compute_next_run` / `_hermes_now` should be scoped the same way (resolve `hermes_time` lazily per tick against the ticked profile's config, not the process's startup context).
2. Cheaper: on tick, when a stored `next_run_at` offset doesn't match the owning profile's configured tz, recompute eagerly (not only when due) — promoting the existing #28934 guard from a due-time repair to a load-time normalization.

## Impact

Silent wrong-time firing for any multi-profile user who runs the desktop app alongside per-profile gateways — every scheduled job fires at (system_tz − profile_tz) offset. Combined with the multiplex ticker, this also means the desktop app's ticks can overwrite correct Eastern `next_run_at` values previously written by the profile gateway, ping-ponging the schedule depending on which process ticked last.

Related: the desktop ticker persistence bug where foreign-profile job sessions land in the desktop's active profile `state.db` (filed separately) — same underlying pattern: the multiplex tick scopes execution per-profile but not all process-level state (timezone, session store, loggers).

SayHell0W0rld:
## Summary

Fixes #97905: a profile's cron jobs had `next_run_at` persisted with UTC offsets despite the profile configured `America/New_York`, causing every job to fire 4 hours early (10:00 ET instead of 14:00 ET for `0 14 * * *`).

## Root Cause

`hermes_time.get_timezone()` caches the resolved zone process-wide (resolved once, cached). The desktop multiplex cron ticker (`_start_desktop_cron_ticker` → `_start_multiplex`) switches profiles via `set_hermes_home_override()` but never resets that cache. So `compute_next_run()` / `mark_job_run()` persist `next_run_at` using the dashboard backend's timezone (often UTC), not the ticked profile's configured timezone. On later ticks the correct profile gateway compares the UTC-stored instant against its Eastern-time `now()`, so jobs fire early.

Verified premise with a repro: switching a fake `HERMES_HOME` override with `timezone: America/New_York` does NOT change `hermes_time.get_timezone()` unless `hermes_time.reset_cache()` is called.

## Approach

Add `hermes_time.reset_cache()` after each `set_hermes_home_override()` in the multiplex ticker's three per-profile loops (recovery, tick, heartbeat). After the override, `hermes_time` re-resolves the timezone from the profile's own `config.yaml` `timezone` key (`get_config_path()` follows `get_hermes_home()` which honors the override).

## Test Plan

- New regression test `test_multiplex_resets_timezone_cache_per_profile`: runs the multiplex ticker over two fake profiles, asserts `hermes_time.reset_cache()` is called per profile switch.
- Full file: 33/33 pass. `ruff` + Windows footgun lint clean.

## Risk / Exclusions

- Only touches the multiplex cron ticker's profile-switch points; non-multiplex cron (single gateway per profile) already runs under the correct `HERMES_HOME` and is unaffected.
- `hermes_time.reset_cache()` is a cheap cache clear; it only forces a config re-read on the next `now()` call.

References #97905

### DIFF
--- a/cron/scheduler_provider.py
+++ b/cron/scheduler_provider.py
@@ -674,6 +674,7 @@ class InProcessCronScheduler(CronScheduler):
             use_cron_store,
         )
         from hermes_constants import set_hermes_home_override, reset_hermes_home_override
+        import hermes_time
 
         logger = logging.getLogger("cron.scheduler_provider")
         logger.info(
@@ -689,6 +690,11 @@ class InProcessCronScheduler(CronScheduler):
         for entry in _existing_profile_homes(profile_homes):
             home = entry[1] if isinstance(entry, tuple) else entry
             home_token = set_hermes_home_override(str(home))
+            # hermes_time caches the resolved timezone process-wide; after
+            # switching to this profile's home, re-resolve so cron instants
+            # persist in the profile's configured zone, not the dashboard
+            # backend's (#97905).
+            hermes_time.reset_cache()
             try:
                 with use_cron_store(home):
                     recovered = self.recover_interrupted()
@@ -713,6 +719,7 @@ class InProcessCronScheduler(CronScheduler):
                     for entry in _existing_profile_homes(profile_homes):
                         home = entry[1] if isinstance(entry, tuple) else entry
                         home_token = set_hermes_home_override(str(home))
+                        hermes_time.reset_cache()
                         try:
                             with use_cron_store(home):
                                 cron_tick(
@@ -736,6 +743,7 @@ class InProcessCronScheduler(CronScheduler):
             for entry in _existing_profile_homes(profile_homes):
                 home = entry[1] if isinstance(entry, tuple) else entry
                 home_token = set_hermes_home_override(str(home))
+                hermes_time.reset_cache()
                 try:
                     with use_cron_store(home):
                         record_ticker_heartbeat(success=ok)
--- a/tests/cron/test_scheduler_provider.py
+++ b/tests/cron/test_scheduler_provider.py
@@ -190,6 +190,52 @@ def test_inprocess_provider_ticks_and_stops():
     assert calls[0].get("sync") is False
 
 
+def test_multiplex_resets_timezone_cache_per_profile():
+    """The multiplex ticker must re-resolve hermes_time after each profile
+    home override; otherwise next_run_at is persisted in the dashboard
+    backend's timezone, not the ticked profile's configured zone (#97905)."""
+    from cron.scheduler_provider import InProcessCronScheduler
+
+    resets = []
+    stop = threading.Event()
+    prov = InProcessCronScheduler()
+    homes = ["C:/fake/home-a", "C:/fake/home-b"]
+
+    def fake_tick(*args, **kwargs):
+        return 0
+
+    def fake_override(home):
+        return "tok"
+
+    def fake_reset_override(token):
+        return None
+
+    with (
+        patch("cron.scheduler.tick", side_effect=fake_tick),
+        patch("hermes_time.reset_cache", side_effect=lambda: resets.append(1)),
+        patch("hermes_constants.set_hermes_home_override", side_effect=fake_override),
+        patch("hermes_constants.reset_hermes_home_override", side_effect=fake_reset_override),
+        patch("cron.jobs.use_cron_store"),
+        patch("cron.jobs.record_ticker_heartbeat"),
+        patch("cron.jobs.clear_ticker_error"),
+        patch("cron.jobs.record_ticker_error"),
+        patch("cron.scheduler_provider._existing_profile_homes", side_effect=lambda h: h),
+    ):
+        t = threading.Thread(
+            target=prov._start_multiplex,
+            args=(stop,),
+            kwargs={"profile_homes": homes, "interval": 0},
+            daemon=True,
+        )
+        t.start()
+        assert _wait_until(lambda: len(resets) >= 2), "multiplex never reset hermes_time cache"
+        stop.set()
+        t.join(timeout=5)
+
+    assert not t.is_alive(), "multiplex did not exit after stop_event was set"
+    assert len(resets) >= 2, "hermes_time.reset_cache must be called per profile switch"
+
+
 # ── Phase 2: config key, discovery, resolver ─────────────────────────────────
 
 


alt-glitch:

Related: #92489 addresses the same multiplexed-profile timezone-cache defect by keying the cache to its active source. This PR resets it at each ticker profile switch; maintainers should choose or combine the approaches.

andrexibiza (commented):
The patch found the correct symptom boundary—the multiplex ticker switches `HERMES_HOME` while `hermes_time` retains process state—and it covers all three profile loops. The fix still does not establish profile ownership of the timezone, though, and the regression never observes the persisted behavior.

### P1 — clearing a process-global cache does not bind it to a context-local profile

[`set_hermes_home_override()` is explicitly `ContextVar`-local](https://github.com/NousResearch/hermes-agent/blob/8416ce494b49838251d43e8eb22fef4dbb383833/hermes_constants.py#L17-L40), while [`hermes_time` stores one unsynchronized process-global `_cached_tz` / `_cached_tz_name` / `_cache_resolved` tuple](https://github.com/NousResearch/hermes-agent/blob/8416ce494b49838251d43e8eb22fef4dbb383833/hermes_time.py#L31-L35) and repopulates it on the next caller of `get_timezone()` ([resolution path](https://github.com/NousResearch/hermes-agent/blob/8416ce494b49838251d43e8eb22fef4dbb383833/hermes_time.py#L96-L121)). The new sequence is therefore only:

1. set profile A in this context;
2. clear shared process state;
3. let whichever thread calls `hermes_time.now()` next repopulate that shared state.

That caller is not guaranteed to be profile A. [Desktop starts the ticker in a daemon thread while the FastAPI/event-loop work continues concurrently](https://github.com/NousResearch/hermes-agent/blob/8416ce494b49838251d43e8eb22fef4dbb383833/hermes_cli/web_server.py#L452-L469), and `tick(sync=False)` intentionally copies the profile `ContextVar` into persistent executor jobs and returns without waiting ([dispatch context and pool submit](https://github.com/NousResearch/hermes-agent/blob/8416ce494b49838251d43e8eb22fef4dbb383833/cron/scheduler.py#L7950-L8030), [async return contract](https://github.com/NousResearch/hermes-agent/blob/8416ce494b49838251d43e8eb22fef4dbb383833/cron/scheduler.py#L8040-L8118)). The multiplex loop can therefore reset the cache for profile B while a profile-A worker is still completing. `mark_job_run()` then takes its `now` from `hermes_time` and overwrites `next_run_at` from that value ([completion timestamp](https://github.com/NousResearch/hermes-agent/blob/8416ce494b49838251d43e8eb22fef4dbb383833/cron/jobs.py#L2750-L2795), [next-run rearm](https://github.com/NousResearch/hermes-agent/blob/8416ce494b49838251d43e8eb22fef4dbb383833/cron/jobs.py#L2860-L2872)). That is the reported corruption path, still reachable without any unrelated caller: A dispatches, the mux advances to B, then A completes under B's shared cache.

There is a second leak in the opposite direction. Each `finally` restores the prior home but does not restore or invalidate the timezone cache ([all three new switch sites](https://github.com/NousResearch/hermes-agent/blob/8416ce494b49838251d43e8eb22fef4dbb383833/cron/scheduler_provider.py#L691-L758)), so after the final profile pass unrelated dashboard work sees the last ticked profile's cached timezone even though its own `ContextVar` points back to the dashboard profile.

This has to be fixed at the ownership layer, not with scheduler call-site resets. Key the cache by the active timezone source/profile identity (or make it context-local), and make resolution/commit atomic so two contexts cannot interleave a mixed `(identity, name, ZoneInfo)` result. Please reconcile that canonical implementation with the existing root-layer work in #92489 and the schedule-zone work in #97491 rather than landing a third overlapping timezone contract without one declared owner.

### P1 — the regression proves only that a helper was called

[`test_multiplex_resets_timezone_cache_per_profile`](https://github.com/NousResearch/hermes-agent/blob/8416ce494b49838251d43e8eb22fef4dbb383833/tests/cron/test_scheduler_provider.py#L193-L238) replaces `cron.scheduler.tick` with a no-op, replaces both profile scopes with no-ops, and counts `reset_cache()` calls. It never loads two profile configs, calls `hermes_time.now()`, runs `compute_next_run()` / `mark_job_run()`, or inspects `jobs.json`. It would remain green if a foreign thread wins the cache fill, if the restored dashboard context inherits the last profile's zone, or if `next_run_at` is still persisted with the wrong offset.

The regression needs to assert the invariant, not the mechanism: two temporary profile homes with different IANA zones; a real recurring job in profile A; an interleaving barrier that lets profile B resolve time before A's async completion; then verify A's persisted `last_run_at` / `next_run_at` retain A's zone and wall-clock schedule, while the restored dashboard context resolves its own zone. That test should fail on current `main` and on this patch, then pass only with profile-owned, concurrency-safe timezone resolution.

### Interlock / exact-head gate

#92489 already changes `hermes_time` cache ownership, and #97491 changes persisted cron timezone semantics. This PR must declare an absorption/supersession path and preserve contributor credit; merging independent partial fixes here would leave order-dependent behavior across three overlapping implementations.

The exact head `8416ce494b49838251d43e8eb22fef4dbb383833` also has no executed green release gate: [CI](https://github.com/NousResearch/hermes-agent/actions/runs/33258946300), [Docker](https://github.com/NousResearch/hermes-agent/actions/runs/33258945990), and [Nix](https://github.com/NousResearch/hermes-agent/actions/runs/33258945903) are all `action_required`, not successful runs. This is not mergeable until the ownership fix, behavioral regression, overlap resolution, and exact-head checks are green.

SayHell0W0rld:
Closing — @andrexibiza's review is correct on both points. (1) reset_cache() clears a process-global while set_hermes_home_override() is ContextVar-local: with tick(sync=False) the mux advances past profile A while A's async completion still reads the shared cache (now repopulated by profile B), so next_run_at can still land in the wrong zone. (2) The regression only counted reset_cache() calls — it never ran compute_next_run/mark_job_run against two real profile configs. The correct fix is ownership-layer (#92489 keys hermes_time's cache to the active profile; #97491 preserves schedule timezone across processes), not scheduler call-site resets. Supporting those; keeping the review queue clean.

PR #54100, opened 2026-06-28
fix(kanban): warn and continue when optional skill is unavailable (#54088)

liuhao1024:
## What does this PR do?

Kanban workers launched by the dispatcher crash with `ValueError("Unknown skill(s): ...")` when a task-level preloaded skill cannot be resolved (e.g. transient skill-index or profile-home differences between the dispatcher and the worker). This blocks an entire swarm stage, including its synthesizer.

This PR detects the Kanban worker context via `HERMES_KANBAN_DB` env var (injected by the dispatcher on every worker spawn) and downgrades the error to a `logger.warning()`, allowing the worker to proceed with whatever skills **are** available.

Direct CLI invocations remain strict — `ValueError` is still raised when the user explicitly requests a skill that doesn't exist.

## Related Issue

Fixes #54088

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `cli.py` — When `missing_skills` is non-empty and `HERMES_KANBAN_DB` is set (Kanban worker context), log a warning and continue with available skills instead of raising `ValueError`. Direct CLI invocations without `HERMES_KANBAN_DB` still raise.
- `tests/cli/test_cli_preloaded_skills.py` — Added `test_main_warns_for_unknown_skill_in_kanban_worker` that sets `HERMES_KANBAN_DB`, provides a mix of available and missing skills, and asserts the worker continues with loaded skills instead of crashing.

## How to Test

1. Run `python -m pytest tests/cli/test_cli_preloaded_skills.py -v` — all 4 tests should pass.
2. Existing test `test_main_raises_for_unknown_preloaded_skill` confirms direct CLI still raises `ValueError` for unknown skills.
3. New test `test_main_warns_for_unknown_skill_in_kanban_worker` confirms Kanban workers warn and continue.

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
- [x] My PR contains **only** changes related to this fix/feature (no unrelated commits)
- [x] I've run `pytest tests/ -q` and all tests pass
- [x] I've added tests for my changes (required for bug fixes, strongly encouraged for features)
[...]

PR #30025, opened 2026-05-21
fix(kanban): gate task.skills on resolvability to prevent worker crash loops

noestelar:
## Problem

The dispatcher already gates the built-in `--skills kanban-worker` injection on resolvability under the worker's `HERMES_HOME` (see `_kanban_worker_skill_available` at `hermes_cli/kanban_db.py:5202`). The reason is documented in that function's docstring: preloading a missing skill is fatal at CLI startup (`ValueError: Unknown skill(s): <name>`), aborting the worker before the agent loop runs.

But per-task `task.skills` entries (`hermes_cli/kanban_db.py:5387-5389` before this patch) were passed through unconditionally:

```python
if task.skills:
    for sk in task.skills:
        if sk and sk != "kanban-worker":
            cmd.extend(["--skills", sk])
```

So a task whose skill name only exists in the global skills root — not the profile-scoped one the worker actually loads — crash-loops the worker until the dispatcher's watchdog eventually auto-blocks it.

## Observed impact

A task with `skills=['mlflow-eval-datasets-for-llm-pipelines']` running under a profile-scoped `HERMES_HOME` respawned **1032 times** before being blocked manually, because the skill lives in `~/.hermes/skills/mlops/` but not in `~/.hermes/profiles/<name>/skills/mlops/`. The CLI fails inside `_apply_skills_overrides` before any tool can call `block`, so the lifecycle's normal error path never runs.

Watchdog auto-block fires far too late (hundreds of respawns) when the failure happens before the agent loop.

## Fix

- Generalize `_kanban_worker_skill_available()` into `_skill_available_for_home(skill_name, hermes_home)`. The legacy name is kept as a back-compat shim that delegates, so external callers (if any) aren't broken.
- Apply the same gate to every `task.skills` entry in `_default_spawn`. Skipped skills emit a single-line stderr warning naming the task id and the missing skill so operators triaging logs see why the worker did not pick up the requested skill. The task still proceeds — running without the supplementary skill context is strictly better than crash-looping until the watchdog gives up.

### Contract change

This is a deliberate **fail-fast → silent-skip-with-warning** shift for `task.skills`. Trade-off:

- Before: dispatcher trusts the caller, missing skill crashes the worker loudly (in the kanban log) but invisibly to the operator (until the auto-block triggers).
- After: dispatcher validates against the actual filesystem the worker will load, drops the unresolvable flag, warns on stderr, and lets the worker run. Operators see the warning in the kanban log alongside the worker's normal output.

[...]

PR #97965, opened 2026-08-29
fix(kanban): make review guidance profile-independent

Rook-CodeVolt:
## Summary

- stop automatic Kanban review dispatch from force-loading the optional bundled `sdlc-review` skill
- carry review-lane identity in task dispatch context so existing two-argument spawn wrappers remain compatible
- inject a compact, session-static review protocol natively for review workers
- preserve explicitly assigned task skills without requiring a bundled skill to exist in the reviewer profile

## Root cause

Profile skill opt-out support predates automatic review dispatch:

- `51f9953e69d303c3d278e41295b1a5c786bf8d87` added `--no-skills` profile creation
- `2ed96372ade3e2f6797b68fb88bf0a53f52f2ee8` made blank-slate profiles durable
- `f55d94a1e0454bc1f2855631d9a0869bc85375dc` later made review dispatch unconditionally append `sdlc-review`

That created a hidden assumption that every reviewer profile can resolve every bundled skill. A valid blank-slate profile instead starts with:

```text
Error: Unknown skill(s): sdlc-review
```

The process exits before its first heartbeat, so ordinary crash reconciliation can only report a dead PID and retries repeat the same deterministic failure.

The repository already established the intended architecture in `84e1d31e5442eeff0bfcf1c2ffab6acf7fe95f45`: mandatory Kanban behavior belongs in native injected guidance, while profile skills remain optional. This change applies that rule to the review lane instead of approximating the runtime skill resolver or silently overriding profile policy.

## Related work

This is narrower than #54100, which makes every Kanban worker continue after missing skill preload, and distinct from #30025/#89601, which filter user-assigned per-task skills at dispatch. This PR preserves strict explicit skill loading and task skill contracts; it removes only the dispatcher's own hidden optional-skill dependency by moving its mandatory protocol to the native prompt path.

## Behavior

Before:

1. a review task is claimed
2. `sdlc-review` is appended to task skills
3. the worker starts with `--skills sdlc-review`
4. a blank-slate reviewer exits before heartbeat

After:

1. review-lane identity is attached to the transient dispatch task
2. legacy spawn wrappers can pass that same task to the default spawner
3. the child receives `HERMES_KANBAN_REVIEW=1`
4. native review guidance is added to the static Kanban system-prompt section
5. no optional skill is force-loaded; explicitly assigned skills are preserved

## Tests

- review dispatch preserves explicit skills and marks the review lane
- default spawn sets review-lane environment without adding `sdlc-review`
- legacy two-argument spawn wrappers preserve review semantics
- ready-lane spawn clears an inherited review marker
- native review guidance is review-only and byte-stable
- focused Kanban, prompt, and bundled-skill suites
- full `tests/hermes_cli/test_kanban*.py` regression surface: `402 passed, 2 skipped`
- Ruff and `git diff --check`

The new regression tests were also run against the pre-fix production path and failed for the expected missing behavior.

## Scope and risk

This does not weaken review verdict handling or independent-review gates. It changes how mandatory review instructions are delivered: native guidance is always available to a review worker, while the larger `sdlc-review` skill remains available when explicitly assigned.

No worker-log persistence, credential handling, board schema, or retry policy is changed.

### DIFF
--- a/agent/agent_init.py
+++ b/agent/agent_init.py
@@ -1632,9 +1632,11 @@ def init_agent(
     # Resolving the ~835-token block once here avoids re-running the
     # membership test + reference on every system-prompt rebuild
     # (init + each context compression).
-    from agent.prompt_builder import KANBAN_GUIDANCE
+    from agent.prompt_builder import kanban_guidance
     agent._kanban_worker_guidance = (
-        KANBAN_GUIDANCE if "kanban_show" in agent.valid_tool_names else ""
+        kanban_guidance(review=os.environ.get("HERMES_KANBAN_REVIEW") == "1")
+        if "kanban_show" in agent.valid_tool_names
+        else ""
     )
 
     # Check tool requirements
--- a/agent/prompt_builder.py
+++ b/agent/prompt_builder.py
@@ -399,6 +399,31 @@ KANBAN_GUIDANCE = (
     "cross-agent handoffs that outlive one API loop."
 )
 
+KANBAN_REVIEW_GUIDANCE = (
+    "# Kanban review-lane protocol\n"
+    "You are the independent reviewer for this task, not its implementer. Treat "
+    "the handoff as a claim to verify, not proof. Read the original task, latest "
+    "handoff, prior requested changes, and actual deliverable. Map every acceptance "
+    "criterion to evidence; for code, inspect the diff and callers, run relevant "
+    "focused tests/lint/type checks, exercise the reported failure and a normal "
+    "control path when practical, and consider error handling, races, data loss, "
+    "security boundaries, and relevant cross-platform behavior.\n"
+    "Choose exactly one terminal verdict. Approve only with concrete evidence via "
+    "`kanban_complete`. For correctable defects, first leave actionable findings "
+    "with locations, reproduction, impact, and minimum correction, then call "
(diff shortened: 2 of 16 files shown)


Rook-CodeVolt:
CI is currently blocked before any jobs start: all three fork-triggered workflows (`CI`, `Docker Build, Test, and Publish`, and `Nix flake check`) completed immediately with `action_required` and no jobs. I confirmed the head is mergeable and attempted the workflow-run approval endpoint, but GitHub correctly requires upstream repository admin rights. Could a maintainer approve the fork workflow runs so CI can execute?

Rook-CodeVolt:
Thanks for triaging this PR, @alt-glitch. The new labels triggered `Label rerun`, but GitHub applies the same first-time fork approval gate to that workflow, so it also completed as `action_required` with zero jobs and cannot bootstrap the original runs.

@teknium1, could you use **Files changed → Awaiting approval → Approve workflows to run**? Exact head `daa93cd79e9e00a41dcfa088cdf8010332795996` is unchanged and mergeable. This PR changes no `.github/workflows/**` files or dependencies; its bounded publication verification was 402 tests passed with 2 platform skips, plus Ruff and independent review. Contributor credentials cannot perform the required upstream Actions approval.

PR #97971, opened 2026-08-29
test(cli): cover SOUL.md legacy-template detection in default_soul

salch-cred:
Add behavioral regression coverage for `hermes_cli.default_soul` — the SOUL.md template-matching module used to safely upgrade legacy comment-only scaffolds to the default persona.

## Why

`default_soul.py` decides whether an old install's `SOUL.md` is safe to replace in place with `DEFAULT_SOUL_MD`. The whole safety guarantee is that a legacy template carries zero user intent — so a bug in `is_legacy_template_soul()` (e.g. a normalization mismatch that wrongly classifies a user-customized SOUL.md as a safe-to-upgrade template) could silently overwrite a user's persona. It had no dedicated unit coverage (only an incidental reference in `test_config.py`).

## What

`tests/hermes_cli/test_default_soul.py` — behavioral tests:

- The default persona is non-empty and carries identity.
- Known legacy templates match verbatim.
- Matching survives trailing newline, CRLF line endings, and a leading UTF-8 BOM (the normalization `_normalize_soul` exists exactly for these).
- A user-authored persona (even one added line) does NOT match — the critical safety property.
- The current `DEFAULT_SOUL_MD` is not a legacy template.
- Degenerate empty/whitespace input does not match.

## Validation

All assertions verified to pass against the current `default_soul.py` on `main`.

### DIFF
--- a/contributors/emails/salch-cred@users.noreply.github.com
+++ b/contributors/emails/salch-cred@users.noreply.github.com
@@ -1,0 +1,1 @@
+salch-cred
--- a/tests/hermes_cli/test_default_soul.py
+++ b/tests/hermes_cli/test_default_soul.py
@@ -1,0 +1,60 @@
+"""Behavioral tests for hermes_cli.default_soul — SOUL.md template matching.
+
+Covers the legacy-template detection used when upgrading an old comment-only
+SOUL.md (seeded by pre-DEFAULT_SOUL_MD installers) to the default persona,
+without touching user-customized content.
+"""
+
+import pytest
+
+from hermes_cli.default_soul import (
+    DEFAULT_SOUL_MD,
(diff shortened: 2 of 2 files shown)


Enough1122:

### PR #97971 — test(cli): cover SOUL.md legacy-template detection in default_soul

**Summary:** Adds `tests/hermes_cli/test_default_soul.py` (8 tests) for `is_legacy_template_soul`: verbatim match, trailing newline, CRLF normalization, BOM stripping, user-persona rejection, empty/whitespace rejection, and that `DEFAULT_SOUL_MD` itself is not considered legacy. Adds contributor email file.

- Title claims tests but also touches `contributors/emails/` — harmless contributor map addition.

**Verdict:** Coverage for installer upgrade safety. No issues.

PR #97977, opened 2026-08-29
fix(memory): single-source the built-in memory cap defaults + log store provenance

smn565s:
## Problem

The built-in memory cap defaults (`2200` / `1375`) are hardcoded independently in **three** places:

- `hermes_cli/config_defaults.py` — `DEFAULT_CONFIG["memory"]`
- `tools/memory_tool.py` — `MemoryStore.__init__` and the `/memory` approval-store fallback builder
- `agent/agent_init.py` — live agent memory-store construction

Because the same literal appears in all three, a config-resolution bug that makes the store fall back to the default is **indistinguishable from a legitimate default** — the failure is completely silent.

### How it bit in practice

On a multiplex gateway (`gateway.multiplex_profiles: true`), the memory store is constructed reading the **raw main `config.yaml`** while `hermes config get` reports the **profile-overlaid** value. Result: the store silently enforced `2200` while the CLI showed the raised cap, with no signal anywhere that the two had diverged. Every memory write hit a wall the operator couldn't see.

## Changes

- Add `DEFAULT_MEMORY_CHAR_LIMIT` / `DEFAULT_USER_CHAR_LIMIT` as the single source of truth in `config_defaults.py` (a pure-data leaf module, no new imports).
- Import them in `memory_tool.py` and `agent_init.py`; delete the three duplicate literals.
- Log store provenance at construction (`memory store: home=… mem_cap=… user_cap=… source=…`) so a profile/main resolution split is diagnosable from one log line instead of being silent.

## Risk

Low. No behavior change to the default values themselves (still `2200` / `1375`). This is pure de-duplication plus one observability log line. Explicit config values and the config-override path are unaffected.

### DIFF
--- a/agent/agent_init.py
+++ b/agent/agent_init.py
@@ -1880,9 +1880,27 @@ def init_agent(
             agent._memory_nudge_interval = int(mem_config.get("nudge_interval", 10))
             if agent._memory_enabled or agent._user_profile_enabled:
                 from tools.memory_tool import MemoryStore
+                from hermes_cli.config_defaults import (
+                    DEFAULT_MEMORY_CHAR_LIMIT,
+                    DEFAULT_USER_CHAR_LIMIT,
+                )
+                _mem_cap = int(mem_config.get("memory_char_limit", DEFAULT_MEMORY_CHAR_LIMIT))
+                _usr_cap = int(mem_config.get("user_char_limit", DEFAULT_USER_CHAR_LIMIT))
+                # Provenance: log the home + caps the store is actually built
+                # against, so a profile/main config-resolution split is
+                # diagnosable from one line instead of silent (#memory-cap-split).
+                try:
+                    from hermes_constants import get_hermes_home
+                    _mem_src = "config" if "memory_char_limit" in mem_config else "default"
+                    logger.info(
+                        "memory store: home=%s mem_cap=%d user_cap=%d source=%s",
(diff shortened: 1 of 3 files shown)


Enough1122:

### PR #97977 — fix(memory): single-source the built-in memory cap defaults + log store provenance

**Summary:** Eliminates three drifting copies of `2200`/`1375` by introducing `DEFAULT_MEMORY_CHAR_LIMIT`/`DEFAULT_USER_CHAR_LIMIT` in `hermes_cli/config_defaults.py` and using them in `DEFAULT_CONFIG`, `tools/memory_tool.py` (`MemoryStore` defaults and `load_on_disk_store`), and `agent/agent_init.py`. `init_agent` now logs `memory store: home=… mem_cap=… user_cap=… source=config|default` via `get_hermes_home()` so profile/main config-resolution splits are diagnosable.

- `logger.info` is wrapped in `try: ... except Exception: pass` so `get_hermes_home()` failures never break agent init — safe.
- `int(mem_config.get(..., DEFAULT_...))` could raise `ValueError` if config stores a non-numeric string; existing code had same risk. Consider a try/except with fallback to default for robustness, but out of scope.
- Default values remain `2200`/`1375` (~800/500 tokens) — no behavioral change.

**Verdict:** Clean single-source deduplication with observability improvement. No blocking issues.

Issue #11237, opened 2026-04-16
[Bug]: `think=False` incorrectly sent to all `provider=custom` endpoints, not just Ollama


wyan-r:
## Summary

When a user sets `reasoning_effort: none` (or disables reasoning via `/reasoning none`),
Hermes Agent sends `extra_body["think"] = false` to **every** custom provider endpoint. The
`think` parameter is Ollama-specific and is rejected with HTTP 422 by any non-Ollama provider
that validates its request body.

## Acceptance Criteria

- [ ] `think=False` is sent to Ollama/local endpoints when reasoning is disabled.
- [ ] `think=False` is **not** sent to cloud custom providers (Mistral, Fireworks, Together.ai, vLLM remote, etc.).
- [ ] OpenRouter and Anthropic reasoning paths are unaffected.
- [ ] `python -m pytest tests/ -q` passes with no regressions.

---

## Labels

`bug` `provider-compat` `reasoning` `ollama`

### Steps to Reproduce

1. Configure any non-Ollama endpoint as a custom provider, e.g. Mistral AI, Fireworks,
   Together.ai, or a vLLM server:
   ```yaml
   provider: custom
   base_url: https://api.mistral.ai/v1
   model: mistral-large-latest
   ```
2. Disable reasoning in the CLI or config:
   ```
   /reasoning none
   ```
   or in `config.yaml`:
   ```yaml
   reasoning_effort: none
   ```
3. Send any message.

### Expected Behavior

The unknown 'think' field should not have been sent to the provider. 

### Actual Behavior

**Result:** The provider returns HTTP 422 because it received an unknown `think` field:

```json
{
  "detail": [
    {
      "type": "extra_forbidden",
      "loc": ["body", "think"],
      "msg": "Extra inputs are not permitted"
    }
  ]
}
```

### Affected Component

Agent Core (conversation loop, context compression, memory)

### Debug Report

```shell
.
```

### Operating System

Linux 6.12.74+deb13+1-amd64 x86_64

### Python Version

3.11.15

### Hermes Version

0.9.0

### Additional Logs / Traceback (optional)

```shell

```

### Root Cause Analysis (optional)

**File:** `run_agent.py` — `_build_api_kwargs()`

The guard condition checks only `provider == "custom"`, which matches all custom providers
regardless of their URL:

```python
if self.provider == "custom" and self.reasoning_config and isinstance(self.reasoning_config, dict):
[...]

PR #64286, opened 2026-07-14
fix(providers): scope think to Ollama endpoints

AXEG0:
## What does this PR do?

Restricts Ollama-only disabled-reasoning controls to endpoints positively identified as Ollama.

When reasoning is disabled, a verified Ollama endpoint receives both `extra_body.think=false` and top-level `reasoning_effort="none"`. Generic OpenAI-compatible custom relays receive neither field, while explicit enabled reasoning effort remains available for custom reasoning APIs. Ollama detection uses the existing `/api/tags` probe instead of URL substrings or port numbers.

## Related Issue

Fixes #11237.

Related/overlapping work:

- #11520 targets the same unsupported-`think` symptom but identifies Ollama from URL text/port. This PR uses positive `/api/tags` detection to avoid misclassifying arbitrary relays.
- #63315 gates enabled reasoning controls on Ollama model capabilities. This PR instead scopes the disabled-control path and protects non-Ollama custom relays.
- #64608 added the current top-level disabled `reasoning_effort` behavior; this PR scopes that field under the same verified-Ollama predicate as `think`.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `plugins/model-providers/custom/__init__.py`: identify Ollama through `detect_local_server_type(base_url, api_key)`; scope both disabled controls to verified Ollama endpoints.
- `agent/chat_completion_helpers.py` and `agent/transports/chat_completions.py`: thread the API key through provider-profile and retry/fallback request construction.
- `agent/auxiliary_client.py`: keep auxiliary request construction in parity with the main transport path.
- Tests under `tests/agent/`, `tests/plugins/model_providers/`, and `tests/providers/`: cover Ollama, misleading URLs/ports, generic relays, API-key propagation, auxiliary requests, and retry/transport parity.

## How to Test

1. Run `scripts/run_tests.sh tests/plugins/model_providers/test_custom_profile.py tests/providers/test_transport_parity.py tests/agent/transports/test_chat_completions.py tests/agent/test_auxiliary_client.py tests/agent/test_probe_cache_followups.py -q`.
[...]

PR #11520, opened 2026-04-17
fix: only send think=false to Ollama, not all custom providers

vominh1919:
## Problem

When `reasoning_effort: none` is set, Hermes sends `extra_body["think"] = false` to **every** custom provider endpoint. The `think` parameter is Ollama-specific and is rejected with HTTP 422 by non-Ollama providers (Mistral, Fireworks, Together.ai, vLLM).

## Fix

Gate `think=false` behind Ollama detection using the same pattern as `_is_ollama_glm_backend()`:
- `self._ollama_num_ctx` is set (auto-detected Ollama)
- OR base_url contains `ollama`
- OR base_url contains `:11434` (Ollama default port)

Fixes #11237

PR #11362, opened 2026-04-17
fix(agent): restrict think=False to Ollama/local endpoints only

yeyitech:
## Summary

- Fixes #11237
- The `think=False` parameter in `extra_body` is Ollama-specific. The previous guard `if self.provider == "custom"` matched **all** custom providers (Mistral, Fireworks, Together.ai, vLLM remote, etc.), causing HTTP 422 errors from cloud APIs that do not recognise the `think` field.
- Replace the guard with an explicit Ollama/local endpoint check that mirrors the pattern already used elsewhere in `run_agent.py` (`_ollama_num_ctx`, `_is_glm_zai_direct`).

## Changes

In `_build_api_kwargs()`, replaced:

```python
if self.provider == "custom" and self.reasoning_config and ...:
```

with:

```python
_is_ollama_endpoint = (
    "ollama" in self._base_url_lower
    or ":11434" in self._base_url_lower
    or is_local_endpoint(self.base_url or "")
)
if _is_ollama_endpoint and self.reasoning_config and ...:
```

`is_local_endpoint` is already imported at the top of `run_agent.py` and `self._base_url_lower` is a cached lowercased copy of the base URL set in the `base_url` property setter.

## Test plan

- [ ] Ollama endpoint (`http://localhost:11434/v1`): `think=False` still sent when reasoning is disabled
- [ ] Ollama via URL keyword (`http://my-ollama-server/v1`): `think=False` still sent
- [ ] Cloud custom provider (e.g. Fireworks, Together.ai): `think=False` no longer sent; no HTTP 422
- [ ] Mistral, vLLM remote: `think=False` no longer sent; no HTTP 422

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #97984, opened 2026-08-29
fix(custom): omit Ollama-only think=false on strict OpenAI-compat endpoints

xxxigm:
## Summary
- Turning reasoning off (`reasoning_effort: none`) was stuffing Ollama's `think: false` into every custom OpenAI-compatible request. Strict hosts like Mistral reject unknown fields (`extra_forbidden` → HTTP 422), so chat dies immediately.
- `think=false` now goes out only for Ollama URLs (port 11434 / `ollama` host). Everyone else still gets top-level `reasoning_effort: "none"` so Ollama `/v1` thinking-off keeps working.
- Fixes the Mistral custom-provider 422 reported against v0.20.6. Same class as #11237; prior PRs (#11296, #11520, #11362, #64286) never landed.

## Test plan
- [x] `scripts/run_tests.sh tests/plugins/model_providers/test_custom_profile.py tests/providers/test_transport_parity.py tests/agent/transports/test_chat_completions.py -q`
- [ ] Custom provider → `https://api.mistral.ai/v1` + Reasoning Off → request succeeds (no `think` in the body)
- [ ] Local Ollama `http://127.0.0.1:11434/v1` + Reasoning Off → still sends `think: false` and `reasoning_effort: "none"`

### DIFF
--- a/plugins/model-providers/custom/__init__.py
+++ b/plugins/model-providers/custom/__init__.py
@@ -6,16 +6,39 @@ Volcengine ARK, vLLM, llama.cpp). Key quirks:
   - ollama_num_ctx → extra_body.options.num_ctx (local context window)
   - reasoning_config disabled → top-level reasoning_effort="none"
     (Ollama /v1/chat/completions ignores think=False — ollama#14820)
-    + extra_body.think = False for /api/chat and proxies
+    + extra_body.think = False only on Ollama URLs (/api/chat and proxies)
   - reasoning_config enabled + effort → top-level reasoning_effort
     (the native OpenAI-compatible format GLM/ARK expect; unset omits it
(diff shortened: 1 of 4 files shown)


kshitijk4poor:
Merged via #98000 — your commits `d02dd32` and `467cdd6` were cherry-picked onto current main with your authorship preserved, with two small follow-ups on top (a `try/except ValueError` guard for malformed ports in the URL heuristic, and test-input alignment). This also fixes #11237.

Right file (the profile plugin, where `think` is emitted post-refactor), right gate (only the field that 422s — `reasoning_effort: "none"` stays unconditional), 3-layer tests. Clean fix — thanks!

PR #97988, opened 2026-08-29
fix(desktop): refresh model picker after gateway reconnect

### LINKED ISSUE #83160: [Bug]: Desktop model picker keeps "Hermes gateway is not connected" after reconnect
## Bug Description

Hermes Desktop's model picker can retain `Hermes gateway is not connected` after the gateway has recovered. In the observed incident, the Desktop status bar already showed `Gateway ready`, but reopening the model picker still displayed the disconnected error and no model catalog.

Clicking `Refresh Models` immediately restored the catalog and allowed the model effort to be changed. No Desktop restart, backend restart, profile switch, or configuration change was needed.

This indicates that the picker is displaying a failed `model.options` query from the earlier disconnected state rather than the gateway's current state.

## Steps to Reproduce

This is an observed reproduction from one incident, with the corresponding query-lifecycle gap verified in both the installed source and current upstream `main`:

1. Open an active session in Hermes Desktop on Windows.
2. During a transient gateway-unavailable interval, open the model picker so its initial `model.options` request fails with `Hermes gateway is not connected`.
3. Wait for the Desktop status bar to show `Gateway ready` again without changing profile or session.
4. Reopen the model picker.
5. Observe that it still shows `Hermes gateway is not connected` instead of loading the catalog.
6. Click `Refresh Models`.
7. Observe that the catalog loads immediately and model/effort selection works.

In the original occurrence, an intentional interruption during preflight context compression preceded the transient unavailable state. That timing is an observed trigger, not a claim that compression is required to reproduce the picker bug.

## Expected Behavior

When the existing gateway transitions back to open, the model picker should invalidate or refetch a previously failed `model.options` query. Once the status bar reports `Gateway ready`, reopening the picker should show the catalog rather than a stale disconnected error.

## Actual Behavior

The status bar reports `Gateway ready`, but the model picker continues to display `Hermes gateway is not connected` and prevents model/effort selection until the user explicitly clicks `Refresh Models`.

## Affected Component

Other — Hermes Desktop / model picker

## Messaging Platform

N/A (Desktop UI)

## Debug Report

No debug bundle was uploaded because it may contain unrelated private session and configuration context. Sanitized local diagnostics:

- Visible live status: `Gateway ready`.
- Simultaneous picker error: `Hermes gateway is not connected`.
- Backend process remained alive and responsive.
- Recovery: one click on `Refresh Models`; no restart, reconnect, profile switch, or configuration change.
- Result: model catalog loaded and the desired effort could be selected.
- No data loss was observed.

The full screenshot is withheld because it contains private session titles, chat text, and a local file path. Those details are unrelated to the failure.

## Operating System

Microsoft Windows 11 Pro, version 10.0.26200 (build 26200)

## Python Version

Hermes runtime: Python 3.11.15

## Hermes Version

Hermes Agent v0.20.0 (2026.8.3), local commit `e8e6ab1fcc657e9a777cf54d9a2088d5917e5448` (+5 carried commits). The installed checkout was 912 commits behind upstream when reproduced.

The live installation was not updated or restarted during incident handling. I separately inspected upstream `main` at `03fa32c92dd445eb64c7f67434dd91b32c40701d`; the same model-options query lifecycle gap remains present there. Runtime reproduction on a build from that exact commit has not been performed.

There was no crash or backend traceback for this UI-state failure. The exact user-visible contradiction and successful manual refresh are included above. Content-bearing logs are intentionally excluded.

## Root Cause Analysis

The frontend query lifecycle explains the observed stale error:

1. [`model-menu-panel.tsx`](https://github.com/NousResearch/hermes-agent/blob/03fa32c92dd445eb64c7f67434dd91b32c40701d/apps/desktop/src/app/shell/model-menu-panel.tsx#L74-L77) queries model options with `modelOptionsQueryKey(profile, activeSessionId)` and calls `requestModelOptions({ gateway, sessionId })`. The query does not subscribe to gateway connection state.
2. [`model-options.ts`](https://github.com/NousResearch/hermes-agent/blob/03fa32c92dd445eb64c7f67434dd91b32c40701d/apps/desktop/src/lib/model-options.ts#L47-L50) keys the query only by `model-options`, profile, and session. A reconnect of the same gateway changes none of those values.
3. When a gateway object exists, [`requestModelOptions`](https://github.com/NousResearch/hermes-agent/blob/03fa32c92dd445eb64c7f67434dd91b32c40701d/apps/desktop/src/lib/model-options.ts#L59-L77) always calls `gateway.request('model.options', ...)`; a request made while that gateway is disconnected rejects with the displayed error.
4. The shared [`QueryClient`](https://github.com/NousResearch/hermes-agent/blob/03fa32c92dd445eb64c7f67434dd91b32c40701d/apps/desktop/src/lib/query-client.ts#L6-L11) has `refetchOnWindowFocus: false`. No connection-state transition is present in the picker query to trigger a fresh request.
5. The explicit refresh path in [`model-menu-panel.tsx`](https://github.com/NousResearch/hermes-agent/blob/03fa32c92dd445eb64c7f67434dd91b32c40701d/apps/desktop/src/app/shell/model-menu-panel.tsx#L84-L105) performs another request and writes the successful result into the same query key. That is exactly what repaired the observed UI.

The status bar reads the live gateway readiness independently, which is why it can correctly show `Gateway ready` while the picker still renders the earlier query failure.

## Proposed Fix

On a gateway transition to `open`, invalidate/refetch the active profile/session's `modelOptionsQueryKey`, and prevent the picker query from running while the gateway is known to be closed. A reconnect should use a normal `model.options` request; it should not need `refresh: true`, because that flag also busts the separate one-hour provider catalog cache.

Add a Desktop regression test that:

1. makes the initial `model.options` query reject as disconnected;
2. transitions the same gateway instance to open;
3. verifies that `model.options` is requested again automatically; and
4. verifies that the catalog replaces the stale error without clicking `Refresh Models`.

## Related issues and PRs (not duplicates)

- [PR #48691](https://github.com/NousResearch/hermes-agent/pull/48691) added the explicit refresh action to bust stale backend provider-catalog data. This report concerns a failed frontend query surviving a gateway reconnect.
- [Issue #59063](https://github.com/NousResearch/hermes-agent/issues/59063) concerns stale custom-provider model ownership after switching providers in Settings.
- [PR #59457](https://github.com/NousResearch/hermes-agent/pull/59457) adds a refresh action to the shared modal picker; it does not invalidate on reconnect.
- [PR #59675](https://github.com/NousResearch/hermes-agent/pull/59675) includes cleanup for actual pooled-backend startup/reconnect failures. In this incident the live gateway was already ready.

AI-assisted investigation disclosure: Codex performed read-only local diagnostics, compared the installed source with current upstream, and searched issues/PRs. The reporter directly observed the UI contradiction and confirmed that `Refresh Models` restored normal operation.

sylvesterkaczmarek:
## What does this PR do?

Fixes the Desktop model picker staying stuck on "Hermes gateway is not connected" after the gateway reconnects.

The model-options query could run while the WebSocket was closed. React Query then retained the rejected request state, while the query itself had no subscription to the gateway's connection lifecycle. This adds a small React bridge for `HermesGateway.onState()` and disables the model-options query while the explicit gateway is down. When the socket returns to `open`, the query becomes enabled again and React Query retries it normally.

The legacy request path where no explicit gateway object is supplied remains enabled.

## Related Issue

Fixes #83160

## Type of Change

## Changes Made

- `apps/desktop/src/app/shell/use-gateway-connected.ts` adds a small hook that mirrors the gateway connection state into React.
- `apps/desktop/src/app/shell/model-menu-panel.tsx` gates the `model.options` query on that connection state so reconnecting produces a clean retry.
- `apps/desktop/src/app/shell/use-gateway-connected.test.tsx` covers closed to open to closed transitions, cleanup, and the legacy no-gateway path.

## How to Test

1. Open Desktop while using a gateway-backed model picker.
2. Interrupt the gateway connection, then allow it to reconnect.
3. Open the model picker and verify the provider/model catalog is restored instead of remaining on the stale "gateway is not connected" error.
4. Run `cd apps/desktop && npx vitest run --project ui src/app/shell/use-gateway-connected.test.tsx`.

## Checklist

- [x] I've added regression coverage
- [ ] Full tests were not run locally; GitHub CI will validate the branch
- [x] Documentation/config/tool-schema changes are N/A
- [x] Cross-platform impact considered: the change only follows the existing gateway connection-state API

### DIFF
--- a/apps/desktop/src/app/shell/model-menu-panel.tsx
+++ b/apps/desktop/src/app/shell/model-menu-panel.tsx
@@ -24,6 +24,7 @@ import { sessionTileDelegate } from '@/store/session-states'
 import type { ModelOptionsResponse } from '@/types/hermes'
 
 import { ModelCatalogMenu, type ModelMenuController } from './model-catalog-menu'
+import { useGatewayConnected } from './use-gateway-connected'
 
 export { ModelMenuCloseContext } from './model-catalog-menu'
 
@@ -53,6 +54,7 @@ export function ModelMenuPanel({ gateway, onSelectModel, profile = 'default', re
   const copy = t.shell.modelMenu
   const [refreshing, setRefreshing] = useState(false)
   const queryClient = useQueryClient()
+  const gatewayConnected = useGatewayConnected(gateway)
   // Bind to THIS surface's SessionView (primary or tile) so each pane's menu
   // shows/switches its own model — not the primary-only globals.
   const view = useSessionView()
@@ -70,8 +72,11 @@ export function ModelMenuPanel({ gateway, onSelectModel, profile = 'default', re
   // dedupes, no second fetch). It must be a live subscription, not a cache
   // peek: with no model in the session store yet, currentPickerSelection falls
   // back to the catalog's reported current, and a non-reactive read would
-  // never repaint that fallback once the catalog resolved.
+  // never repaint that fallback once the catalog resolved. Keep the query
+  // disabled while an explicit gateway socket is down so a reconnect toggles
+  // it back on and gives React Query a clean retry boundary (#83160).
   const modelOptions = useQuery({
+    enabled: gatewayConnected,
     queryKey: modelOptionsQueryKey(profile, activeSessionId),
     queryFn: (): Promise<ModelOptionsResponse> =>
       requestModelOptions({ gateway, profile, request: requestGateway, sessionId: activeSessionId })
--- a/apps/desktop/src/app/shell/use-gateway-connected.test.tsx
+++ b/apps/desktop/src/app/shell/use-gateway-connected.test.tsx
@@ -1,0 +1,52 @@
+// @vitest-environment jsdom
+import { act, renderHook } from '@testing-library/react'
+import { describe, expect, it, vi } from 'vitest'
+
+import type { HermesGateway } from '@/hermes'
+
+import { useGatewayConnected } from './use-gateway-connected'
+
+describe('useGatewayConnected', () => {
+  it('tracks closed and reopened gateway states', () => {
+    let state: 'closed' | 'open' = 'closed'
+    let listener: ((next: 'closed' | 'open') => void) | undefined
+    const unsubscribe = vi.fn()
+
+    const gateway = {
+      get connectionState() {
+        return state
+      },
+      onState(next: (value: 'closed' | 'open') => void) {
+        listener = next
+        next(state)
+
+        return unsubscribe
+      }
+    } as unknown as HermesGateway
+
+    const { result, unmount } = renderHook(() => useGatewayConnected(gateway))
+
+    expect(result.current).toBe(false)
+
+    act(() => {
+      state = 'open'
+      listener?.('open')
+    })
+    expect(result.current).toBe(true)
+
+    act(() => {
+      state = 'closed'
+      listener?.('closed')
+    })
+    expect(result.current).toBe(false)
+
+    unmount()
+    expect(unsubscribe).toHaveBeenCalledOnce()
+  })
+
+  it('treats the legacy request path without an explicit gateway as connected', () => {
+    const { result } = renderHook(() => useGatewayConnected(undefined))
+
+    expect(result.current).toBe(true)
+  })
+})
--- a/apps/desktop/src/app/shell/use-gateway-connected.ts
+++ b/apps/desktop/src/app/shell/use-gateway-connected.ts
@@ -1,0 +1,28 @@
+import { useEffect, useState } from 'react'
+
+import type { HermesGateway } from '@/hermes'
+
+/**
+ * React bridge for the gateway's imperative connection-state emitter.
+ *
+ * Model catalog queries must not run while the socket is down: a rejected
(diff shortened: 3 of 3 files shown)


Enough1122:

### PR #97988 — fix(desktop): refresh model picker after gateway reconnect

**Summary:** Adds `useGatewayConnected(gateway)` hook bridging `HermesGateway.onState` to React. `ModelMenuPanel`'s `useQuery` for `modelOptions` is now `enabled: gatewayConnected`, so a failed `model.options` request while the socket is down isn't cached as an error and the reconnect toggles the query back on giving React Query a clean retry boundary (#83160). Hook treats legacy `gateway===undefined` as connected. Tests cover closed→open→closed transitions and unsubscribe.

- `useEffect` dependency on `gateway` object identity may re-subscribe if gateway is recreated; `onState` returns unsubscribe so no leak, but `setConnected` will be called synchronously with initial `gateway.connectionState` via next render — current impl defers via callback only, initial state already set via `useState(() => ...)`. Minor double-set but harmless.
- Disabling query while disconnected means the picker will show loading/previous data rather than stale error — intended.

**Verdict:** Correct React Query retry-boundary fix. No issues.

Issue #97948, opened 2026-08-29
Manual /compress reports 120s timeout while background worker succeeds minutes later; large-session compression fails with lease lost / session_split_failed


AsphaltUranium:
## Summary

Two distinct context-compression problems on large sessions (both observed on Hermes v0.20.6, Windows 11, desktop app):

**A. Manual `/compress` reports a 120s timeout while the background worker actually succeeds minutes later** — silently rotating the session to a new ID. The UI tells the user "timeout" but the compression completes in the background, leaving the user unsure whether it worked.

**B. Automatic (preflight) compression on large sessions repeatedly fails to finish** — the user's message waits ~600s (the ceiling) with the context never reduced, the compression aborts/rolls back, and the next turn re-triggers the same doomed compression. On a ~486k-token session the transcript stays unchanged (1497->1497) after 11+ minutes of work.

## Environment

- Hermes v0.20.6, Windows 11, desktop app (`platform=desktop`)
- provider=coing (Volcano Ark) / model=deepseek-v4-flash-ga-260731, context=1,000,000
- Affected sessions: ~200k–486k request tokens

## Symptom A — Manual /compress: 120s UI timeout, background success

User runs manual `/compress` on a ~200k-token session:

- UI shows a **120s timeout** error.
- But compression **keeps running in the background** and **succeeds ~6.4 minutes later**:
  - `context compression done: session=... messages=330->198 rough_tokens=~208,780`
  - `commit_status=committed`, `total_duration_ms=383235` (6.4 min)
- The session is silently rotated to a new ID (`...69d42f`), so the user must rediscover that it actually worked.

The **frontend wait timeout (120s) is far shorter than the real compression time (6+ min)** — the UI reports failure for work that succeeds. There is no "compression still running in background" state and no later reconciliation of the rotated session ID.

## Symptom B — Automatic compression: never finishes, context stays maxed

On a ~486k-token session (1497 messages), automatic pre-API compression:

1. Triggers at threshold (500k), streams, but **does not finish within the 600s ceiling**:
   - `Context compression still streaming after 120s (last progress 0.0s ago) — extending wait (ceiling 600s)`
   - The user's turn waits the full 600s, then proceeds without compression.
2. The user's next message **re-triggers the same compression** on the unchanged ~486k context — the same hang repeats, so every message costs ~10 minutes.
3. When it does eventually finish the summary, publication fails:
   - `Compression rotation aborted and rolled back to the parent session (...): Compression lease lost before publication`
[...]

PR #97989, opened 2026-08-29
fix(agent): fence compression timeout recovery

adzer000:
## Summary
- bound compression timeout/cancellation effects to attempt, generation, and turn ownership
- prevent stale workers, late callbacks, and superseded retries from publishing into newer turns
- serialize outbound dispatch admission without holding coordination locks across RPCs
- preserve non-cooperative admitted work while returning bounded terminal outcomes
- suppress same-turn automatic retries after timeout while allowing explicit forced retries

## Verification
- canonical focused runner: **235 passed, 0 failed**
- independent Sol review: **APPROVE**, 0 Critical / 0 Important
- `py_compile`: all 10 changed Python paths
- `git diff --check`: pass
- rebased by clean cherry-pick onto current upstream main `3f36c87e1ebdfbf7d14a88229dc9be222c12ea89`

## Runtime
No live Hermes/Desktop restart or activation was performed.

### DIFF
--- a/agent/auxiliary_client.py
+++ b/agent/auxiliary_client.py
@@ -325,6 +325,93 @@ class AuxiliaryExplicitCancellation(BaseException):
         super().__init__("auxiliary request explicitly cancelled by host")
 
 
+class AuxiliaryCancellationToken:
+    """Monotonic attempt-local cancellation and dispatch admission fence."""
+
+    def __init__(self, source_check: Callable[[], Any]) -> None:
+        self._source_check = source_check
+        self._lock = threading.Lock()
+        self._state = "OPEN"
+        self._cancelled = False
+        self._callbacks: list[Callable[[], None]] = []
+
+    def _source_requested(self) -> bool:
+        try:
+            return bool(self._source_check())
+        except Exception:
+            logger.debug("auxiliary cancellation source check failed", exc_info=True)
+            return False
+
+    def _observe_locked(self) -> list[Callable[[], None]]:
+        if not self._cancelled and self._source_requested():
+            self._cancelled = True
+            if self._state == "OPEN":
+                self._state = "CANCELLED"
+            callbacks = self._callbacks
+            self._callbacks = []
+            return callbacks
+        return []
+
+    @staticmethod
+    def _run_callbacks(callbacks: list[Callable[[], None]]) -> None:
+        for callback in callbacks:
+            try:
+                callback()
+            except Exception:
+                logger.debug("auxiliary cancellation cleanup failed", exc_info=True)
+
+    def is_cancelled(self) -> bool:
+        with self._lock:
+            callbacks = self._observe_locked()
+            cancelled = self._cancelled
+        self._run_callbacks(callbacks)
+        return cancelled
+
+    def raise_if_cancelled(self) -> None:
+        if self.is_cancelled():
+            raise AuxiliaryExplicitCancellation()
+
+    def cancel(self) -> None:
+        with self._lock:
+            self._cancelled = True
+            if self._state == "OPEN":
+                self._state = "CANCELLED"
+            callbacks = self._callbacks
+            self._callbacks = []
+        self._run_callbacks(callbacks)
+
+    def claim_dispatch(self) -> None:
+        """Linearize OPEN -> ADMITTED without holding the lock over RPC."""
+        with self._lock:
+            callbacks = self._observe_locked()
+            admitted = not self._cancelled and self._state != "CANCELLED"
+            if admitted:
+                self._state = "ADMITTED"
+        self._run_callbacks(callbacks)
+        if not admitted:
+            raise AuxiliaryExplicitCancellation()
+
+    def register_cancel_callback(self, callback: Callable[[], None]) -> None:
+        with self._lock:
+            callbacks = self._observe_locked()
+            if not self._cancelled:
+                self._callbacks.append(callback)
+                callbacks = []
+            else:
+                # Registration can lose the acquire -> register race after
+                # cancellation has already won. Deliver the late callback
+                # immediately, but only after leaving the token lock below.
+                callbacks.append(callback)
+        self._run_callbacks(callbacks)
+
+    def unregister_cancel_callback(self, callback: Callable[[], None]) -> None:
+        with self._lock:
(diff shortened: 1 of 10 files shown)


andrexibiza (commented):
I reviewed exact head `8a7ee25232147627deffcad5acc70a0efe384e89` against its parent/base `3f36c87e1ebdfbf7d14a88229dc9be222c12ea89`, the current compression/cancellation implementation, the new race tests, and the overlapping compression work already in flight. The direction is strong: one attempt-local cancellation token, a physical-dispatch admission seam, fence-bound terminal identity, and future-owned compression admission are all materially better than re-reading mutable host state after timeout.

I still see two P1 ownership holes before this is safe to land.

### P1 — cancellation returns the *physical RPC* permit while that RPC can still be alive

`call_llm()` now registers `_release_permit_once` as an `AuxiliaryCancellationToken` callback. Once the token cancels, that callback releases the task `BoundedSemaphore` immediately. Separately, `_run_protected_sync_provider_call()` is explicitly allowed to return `AuxiliaryExplicitCancellation` to the owning compression worker while the daemon provider callback remains blocked/non-cooperative.

The new test `test_host_timeout_releases_compression_pool_slot_before_provider_returns` pins the other half of the same state: after host timeout, the compression admission count is already `0` while the first provider callback is still blocked, and a fresh compression is admitted. Put those two mechanisms together and there is no remaining lease on the actual side effect:

1. physical provider RPC A is admitted and blocks;
2. host timeout cancels the token;
3. the auxiliary semaphore is returned by the cancellation callback;
4. the protected wrapper unwinds, so the compression future finishes and returns its compression admission slot;
5. a later turn can admit compression B and physically dispatch RPC B while A is still alive.

The same-turn terminal latch reduces one retry path, but it does not bound this across later turns or another attempt against the unchanged source. A repeatedly hung provider can therefore accumulate detached physical RPCs even though both logical admission layers report capacity. That is the other side of the exact class described in #97648: commit authority is revoked, but provider work can outlive ownership and multiply.

**Required fix:** distinguish logical-attempt cancellation from physical-RPC accounting. The task concurrency permit (or a separate physical-dispatch lease) must remain owned until the real provider callback settles, even if the caller gets its bounded cancelled outcome earlier. Cancellation should deny every *future* dispatch, but it must not advertise physical capacity that does not exist. A deterministic regression should hold `max_concurrency=1`, block provider A after dispatch, timeout/cancel its owner, then start a second-turn compression and prove provider B does **not** execute until A actually returns. If keeping the compression slot free is important for liveness, that is fine; the provider-side lease is the boundary that must remain occupied.

### P1 — `host_timeout_won()` disables the whole same-attempt rollback, not just stale cooldown rollback

`_restore_compressor_attempt_state(..., attempt_fence=...)` now immediately returns whenever `attempt_fence.host_timeout_won()` is true. That solves the narrow test in `test_host_timeout_wins_late_same_generation_cooldown_rollback`: the host's newly written timeout cooldown is not rolled back. But the snapshot is much broader than cooldown state. It also owns speculative attempt-local fields including `_previous_summary`, `compression_count`, summary/fallback state, anti-thrash state, and telemetry.

That conflicts with the ownership contract already merged in #96949: **generation owns compressor-attribute writes; stale-generation rollback is suppressed, but same-attempt rollback is preserved.** There is a real race here. `compress_fn` can finish and mutate compressor bookkeeping, then the host timeout can win before `begin_commit`; the worker sees `_cancel_source()`, raises `AuxiliaryExplicitCancellation`, calls `_restore_compressor_attempt_state()`, and the new early return leaves all of those uncommitted mutations live. The session/transcript did not commit, but the shared compressor can now carry summary/bookkeeping from a result that never became authoritative.

**Required fix:** do not skip the entire rollback when host timeout owns settlement. Restore the same-attempt speculative snapshot while preserving only the fields the host timeout legitimately superseded (terminal outcome + timeout cooldown/durable cooldown row, and any explicitly host-owned retry latch). A regression should snapshot `_previous_summary`/`compression_count`, mutate them as a completed-but-uncommitted compression would, claim the host timeout, run the rollback, and assert the speculative fields return to the snapshot while the timeout terminal record/cooldown remain intact.

### Interlock / merge-order

This head is not an isolated compression fix. The current graph matters:

- **#96949 is merged foundation**, not superseded history. Its generation-vs-commit ownership split and same-attempt rollback contract need to survive this PR.
- **#97064** is open and narrower on fence cancellation reaching the streaming summary; this PR overlaps that mechanism in `agent/conversation_compression.py` and appears capable of superseding that slice once the physical-RPC ownership hole above is closed.
- **#97512** is open and collides in `agent/context_compressor.py`, `agent/conversation_compression.py`, and several of the same tests. It also owns total-deadline/lease semantics. These should not land independently without an explicit winner/rebase order.
- **#97671** is open and overlaps both production files plus `test_auxiliary_explicit_cancellation.py`, but it is **not** a pure duplicate: its attempt-scoped eight-call budget is absent here. Preserve that contribution/requirement even if this PR becomes the cancellation superseder.
- **#97948** remains broader than this diff. In particular, the outer Desktop/compute-host 120s waiter losing the late settlement and the lease-loss/rotation reconciliation path are outside these ten files. This PR should not be treated as closing that issue.

If this PR is intended as the consolidation vehicle, please say that explicitly in the body and preserve the narrower contributor lineage rather than leaving #97064/#97512/#97671 looking like unrelated duplicate work.

### Exact-object verification

The PR is one commit on `3f36c87e`; current `main` is now `1e21fe862421081e14af4564076af74bc58c050b`, four commits beyond that base. More importantly, exact-head CI is not a landing receipt yet: CI run `33260992857`, Nix `33260992599`, and Docker `33260992502` are all `action_required`, and each currently has **0 jobs**. The reported 235 focused local passes are useful development evidence, but they do not make this exact commit green upstream.

There is a lot worth keeping here. In particular, the pre-dispatch fence closes several retry seams cleanly, the stale-turn terminal-publication test is good, and making compression admission future-owned fixes a real host-unwind accounting bug. The remaining blockers are both ownership-boundary bugs: one releases authority/capacity too early at the provider edge, and one preserves speculative state too broadly at settlement. Fix those without losing the existing generation and call-budget work, then recompose on current main and get executable exact-head CI.

PR #63070, opened 2026-07-12
fix(kanban): honor configured kanban.failure_limit on crash and runtime-timeout paths

james47kjv:
### Problem
`detect_crashed_workers` and `enforce_max_runtime` call `_record_task_failure` with `failure_limit=1 if (protocol_violation or is_systemic) else None`; `None` falls to `DEFAULT_FAILURE_LIMIT=2`. Only `recompute_ready`/`dispatch_once` receive the configured `kanban.failure_limit`. Result: ordinary PID-death crashes and runtime timeouts block cards at 2 regardless of an operator's configured limit (e.g. 4), and on restart waves this produces `gave_up`→`promoted` events ~1s apart.

### Fix
Thread the already-loaded `kanban.failure_limit` (the gateway reads it and passes it to `dispatch_once`) into `detect_crashed_workers` and `enforce_max_runtime` too. The deliberate anti-loop override — protocol-violation OR systemic (≥3 same-fingerprint crashes in one sweep) forces limit 1 — is preserved verbatim. New params are keyword-only with `None` default, so all existing callers are unchanged.

### Tests
`tests/hermes_cli/test_kanban_db.py` +3: a plain crash with `failure_limit=4` stays `ready` through crashes #1–3 and blocks only on #4; a systemic burst still trips at the first sweep despite limit 4; the runtime-timeout path honors config too. Full `test_kanban_db.py` + `test_kanban_core_functionality.py` = 396 passed, no regressions.

PR #97991, opened 2026-08-29
fix(kanban): honor dispatcher failure limit in reapers

luken:
## Summary

- propagate the dispatcher `failure_limit` into max-runtime timeout accounting
- propagate the same limit into ordinary worker-crash accounting
- preserve the dedicated systemic-crash and protocol-violation thresholds
- cover strict and lenient limits through full dispatcher ticks

## Bug

The dispatcher passed its configured failure limit to spawn-failure accounting and `recompute_ready`, but the timeout and crash reapers omitted it. Those paths silently fell back to `DEFAULT_FAILURE_LIMIT`.

With a strict limit, tasks received more retries than configured. With a lenient limit, the breaker could trip at the built-in default and `recompute_ready` could immediately promote the task under the configured limit, producing a block/promote retry loop.

## Validation

- reproduced both affected paths on upstream `main` before the fix
- `scripts/run_tests.sh tests/hermes_cli/test_kanban*.py -q`
- 316 passed, 2 platform-specific skips on macOS
- adversarial Claude Code review converged after four bounded passes; unsafe real-signal behavior in the dispatcher-level timeout test was removed before submission

The production change is keyword-only and default-preserving: the dispatcher default and direct-call fallback both remain `DEFAULT_FAILURE_LIMIT`.

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -4519,7 +4519,7 @@ def _resume_status_from_events(conn: sqlite3.Connection, task_id: str) -> str:
 
 
 def recompute_ready(
-    conn: sqlite3.Connection, failure_limit: int = None,
+    conn: sqlite3.Connection, failure_limit: Optional[int] = None,
 ) -> int:
     """Promote ``todo`` tasks to ``ready`` when all parents are ``done`` or ``archived``.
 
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: #63070 previously addressed the same configured failure-limit propagation for crash and timeout reapers. That PR is closed and unmerged; this active reimplementation should be evaluated on its own.

PR #96685, opened 2026-08-27
fix(linux): resolve venv symlink chain when generating desktop entry

aardbij26-ship-it:
## Problem

On Linux installs using a uv-managed venv, the generated desktop entry has a non-runnable `Exec` line, so launching Hermes from the dock/menu icon silently fails. Fixes #90292.

## Root cause

uv venvs chain their interpreter symlinks: `venv/bin/python3 -> python -> uv-managed python`. `_needs_interpreter()` compared the shebang path against `sys.executable`'s directory as **raw strings**, so the uv-managed interpreter was not recognized as the running interpreter and got prefixed explicitly:

```
Exec=/home/user/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/bin/python3.11 /home/user/.hermes/hermes-agent/venv/bin/hermes desktop
```

That interpreter runs **without venv activation** and dies on `ModuleNotFoundError: No module named 'hermes_cli'` — silently, since `Terminal=false`.

## Fix

Compare **resolved** paths instead of raw strings, so the venv symlink chain is recognized as the running interpreter and the entry stays a plain `hermes desktop` launch.

## Test

Adds `test_exec_leaves_symlinked_venv_shebang_alone` covering the uv-venv symlink chain case. Full suite: 16 passed, 2 skipped.

Issue #94110, opened 2026-08-24
[Bug]: Clicking the Linux desktop icon silently fails to launch when the Hermes venv Python is a symlink into the uv-managed interpreter store


Dylan-v5:
**Bug Description**

On Linux, clicking the Hermes icon in the app menu / dock does nothing — no window, no error dialog — while `hermes desktop` from a terminal works perfectly. The auto-generated desktop entry (`~/.local/share/applications/hermes.desktop`) contains an `Exec` line that runs the venv's console script with a bare, non-venv Python interpreter, which cannot see the venv's `site-packages`, so the process dies instantly with `ModuleNotFoundError: No module named 'hermes_cli'` — silently, because `Terminal=false`.

The root cause is a symlink-resolution bug in the desktop-entry generator (details below). This affects anyone whose venv Python is a symlink into the uv-managed interpreter store — which is the default for uv-created venvs, including the ones created by the official installer.

**Steps to Reproduce**

1. Install Hermes on Linux with a uv-managed venv (e.g. the official shell installer, or `uv venv` + git checkout). Confirm `venv/bin/python3` is a symlink into `~/.local/share/uv/python/<version>/`.
2. Run `hermes desktop` once from a terminal — this registers the launcher entry.
3. Quit that instance (or close the terminal), then click the Hermes icon in the app menu / dock.
4. Nothing happens. No window, no error.

Running the generated `Exec` manually confirms the failure:

```bash
$ ~/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/bin/python3.11 ~/.hermes/hermes-agent/venv/bin/hermes desktop
Traceback (most recent call last):
  File "/home/<user>/.hermes/hermes-agent/venv/bin/hermes", line 4, in <module>
    from hermes_cli.main import main
ModuleNotFoundError: No module named 'hermes_cli'
```

**Expected Behavior**

Clicking the launcher icon opens the Hermes Desktop app, same as `hermes desktop` from a terminal, independent of any terminal session.

**Actual Behavior**

The generated `Exec=` line in `~/.local/share/applications/hermes.desktop` is:

```
Exec=/home/<user>/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/bin/python3.11 /home/<user>/.hermes/hermes-agent/venv/bin/hermes desktop
```

[...]

Issue #94058, opened 2026-08-24
Linux desktop entry Exec resolves venv symlink to bare interpreter, breaking launcher launch after upgrade


GitTradWang:
## Summary
On uv-created venvs (and any venv where `bin/python` is a symlink into a shared base-interpreter tree), `hermes desktop` and `hermes update` write a broken `Exec=` line into `~/.local/share/applications/hermes.desktop`. Launching Hermes from the launcher icon then crashes instantly and silently (`Terminal=false`).

## Root cause
`hermes_cli/linux_desktop_entry.py` → `resolve_exec_command()` calls `Path(sys.executable).resolve()`. `uv venv` (and pip) make `bin/python` a symlink into a shared base interpreter tree (e.g. `~/.local/share/uv/python/cpython-3.11.15-.../bin/python3.11`); `.resolve()` follows that link. The generated `Exec` therefore prefixes the *bare* base interpreter, which never activates the venv — CPython discovers `pyvenv.cfg` from the *unresolved* argv[0] — so the app dies on the first third-party import (`ModuleNotFoundError: yaml`).

Because every `hermes desktop` re-installs the desktop entry and `hermes update` rebuilds it via `python -m hermes_cli.main desktop --build-only`, the icon entry is re-broken after every upgrade, while `hermes desktop` from a shell keeps working.

## Repro
1. `uv venv` (or any venv where `bin/python` is a symlink)
2. `hermes desktop` once
3. Inspect `~/.local/share/applications/hermes.desktop` — `Exec` starts with the resolved base interpreter
4. Launch from the launcher icon → instant silent crash

## Environment
- Fedora Linux x86_64, uv-managed venv, Python 3.11

Fix submitted in #94051.

Javierkaiser:
### Independent confirmation (Fedora 44 · GNOME · uv-managed venv)

Hit this exact failure today on a fresh install — confirming the root-cause analysis above, including the *re-broken after every upgrade* part, which makes this worse than a one-off.

**Environment**

- Hermes Agent @ `41447a6d7063` (main), install method: shell installer (git checkout at `~/.hermes/hermes-agent`)
- OS: Fedora 44, GNOME (Wayland)
- Python: uv-managed cpython 3.11.16; venv at `~/.hermes/hermes-agent/venv` where `bin/python` is a symlink into `~/.local/share/uv/python/cpython-3.11.16-linux-x86_64-gnu/bin/python3.11`

**Generated entry (broken):**

```ini
[...]

PR #94051, opened 2026-08-24
fix: keep venv interpreter path in desktop entry Exec

GitTradWang:
## What does this PR do?

Fixes the Linux desktop-entry `Exec` line being rewritten with a bare (non-venv) Python interpreter, which made launching Hermes Desktop from the launcher icon crash **silently** after every upgrade on uv-created venvs.

**Root cause:** `resolve_exec_command()` in `hermes_cli/linux_desktop_entry.py` called `Path(sys.executable).resolve()`. `uv venv` (and pip) make `bin/python` a symlink into a shared base-interpreter tree, and `.resolve()` follows that link — so the generated `Exec` prefixed the *bare* base interpreter (e.g. `~/.local/share/uv/python/.../bin/python3.11`). That interpreter never activates the venv (CPython discovers `pyvenv.cfg` from the *unresolved* argv[0]), so it crashes on the first third-party import (`ModuleNotFoundError: yaml`), with nothing on screen because `Terminal=false`.

Every `hermes desktop` run re-installs the desktop entry (`main.py` → `_register_linux_desktop_entry()`), and `hermes update` rebuilds it through `python -m hermes_cli.main desktop --build-only`. So after each upgrade the icon entry was broken again, while `hermes desktop` from a shell kept working — confusing: "works in terminal, flashes and dies from the icon".

**Fix:**
- Add `_resolve_exec_path()`: when the executable sits inside a venv (a `pyvenv.cfg` exists above it), keep the *unresolved* venv path; resolve only otherwise. Use it for both `sys.executable` and the resolved hermes binary, so the `Exec` prefix stays on the venv interpreter that actually loads site-packages.
- Add `_is_python_interpreter()`: when the resolved bin is the interpreter itself (the updater drives `python -m hermes_cli.main desktop --build-only`, so `argv[0]` is the interpreter), emit `<venv-python> -m hermes_cli.main desktop` instead of `<venv-python> desktop` (which would treat `desktop` as a script path).

## Related Issue

Fixes #94058

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `hermes_cli/linux_desktop_entry.py`:
[...]

PR #97992, opened 2026-08-29
fix(cli): don't resolve venv python into Linux desktop Exec=

flowersjus:
## Bug Description

After `hermes desktop` (and after every rebuild/update), the Linux menu entry is rewritten with an `Exec=` that launches **system Python**. The icon does nothing. `hermes desktop` from a TTY still works.

Reproduced on EndeavourOS / Hyprland, git install, `venv/bin/python -> /usr/bin/python3.11`:

```
Exec=/usr/bin/python3.11 ~/.hermes/hermes-agent/venv/bin/hermes desktop
```

That command exits 1 with `ModuleNotFoundError: No module named 'hermes_cli'`. `Terminal=false`, so the DE shows nothing.

Related: #90292 (partially addressed by #90492), #92095, #94110, #94058.

There are already open PRs on the same class of bug (#92090, #94051, #96685). This one is the smallest delta that also closes a `#90292` hole the others still have: substring-matching the interpreter dir against the shebang, so `/usr/bin` matches `#!/usr/bin/env python3` and skips the venv prefix.

## Root Cause

1. `Path(sys.executable).resolve()` follows `venv/bin/python` (symlink) to `/usr/bin/python3.N` or the uv store. CPython only activates a venv via the *unresolved* argv[0] + `pyvenv.cfg`.
2. `_needs_interpreter` used `exe_dir not in shebang`. After (1), `exe_dir` is `/usr/bin`, which is **not** in `#!…/venv/bin/python3`, so a *correct* venv console-script was classified as foreign and got the system interpreter prefixed. The same substring also matches `#!/usr/bin/env python3`, which would skip the #90292 prefix.

## Fix

- Prefix `sys.executable` **without** following the symlink.
- Treat `env` shebangs as always needing that prefix.
- Compare the unresolved interpreter dir against the shebang *program*, not the whole line.

## How to Verify

1. Linux git/uv install where `venv/bin/python` is a symlink to a base interpreter.
2. Run `hermes desktop` once from a TTY.
3. `awk -F= '/^Exec=/{print substr($0,6)}' ~/.local/share/applications/hermes.desktop`
   - must **not** start with `/usr/bin/python` or `~/.local/share/uv/python/…`
   - good: `…/venv/bin/hermes desktop` or `~/.local/bin/hermes desktop` or `…/venv/bin/python …/hermes desktop`
4. Launch from the app menu. Window opens.

## Test Plan

- [x] Added regression test: venv python symlink must not appear as the resolved base interpreter in `Exec=`
- [x] Existing tests still pass (`tests/hermes_cli/test_linux_desktop_entry.py`: 16 passed, 2 skipped)
- [x] Manual verification of the fix (menu Exec rewritten; `~/.local/bin/hermes --version` works)

## Risk Assessment

Low — Linux `.desktop` writer only. macOS/Windows unchanged. Shell-wrapper launchers still left alone. `#90292` env-shebang prefix still applied, now with the venv path instead of the base interpreter.

### DIFF
--- a/hermes_cli/linux_desktop_entry.py
+++ b/hermes_cli/linux_desktop_entry.py
@@ -76,14 +76,20 @@ def resolve_exec_command() -> str:
             # installer's bash wrapper). Launched from the .desktop entry that
             # shebang resolves to the SYSTEM python and dies on the first
             # third-party import (#90292) — silently, since Terminal=false.
-            # sys.executable is the interpreter actually running Hermes (the
-            # venv one), so prefix it explicitly.
-            argv = [str(Path(sys.executable).resolve()), str(resolved), "desktop"]
+            # Prefix the running interpreter. Do NOT Path.resolve() it:
+            # venv/bin/python is often a symlink to /usr/bin/python3.N, and
+            # following that drops the venv (ModuleNotFoundError: hermes_cli).
+            argv = [_running_interpreter(), str(resolved), "desktop"]
         else:
             argv = [str(resolved), "desktop"]
     else:
-        argv = [str(Path(sys.executable).resolve()), "-m", "hermes_cli.main", "desktop"]
+        argv = [_running_interpreter(), "-m", "hermes_cli.main", "desktop"]
     return " ".join(_quote_exec_arg(a) for a in argv)
+
+
+def _running_interpreter() -> str:
+    """``sys.executable`` as the DE should invoke it — never follow venv symlinks."""
+    return str(Path(sys.executable))
 
 
 def _needs_interpreter(bin_path: Path) -> bool:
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: #92090, #92516, and #94115 address the same Linux desktop Exec= / symlinked-venv family. This PR uses program-level shebang parsing and treats env shebangs as requiring the venv prefix; maintainers should select a consolidated repair.

hermes_cli/linux_desktop_entry.py
```diff
@@ -99,15 +105,24 @@ def _needs_interpreter(bin_path: Path) -> bool:
...
+    if "python" not in rest:
         # A shell wrapper (e.g. the installer's bash launcher) execs the venv
         # python itself — leave it alone.
         return False
+    prog = rest.split()[0]
+    # ``#!/usr/bin/env python3`` always escapes the venv when the DE spawns
+    # it. Do not substring-match the interpreter dir against the shebang:
+    # ``/usr/bin`` is inside ``/usr/bin/env python3`` and would skip the
+    # prefix (#90292) *and* is the follow-symlink target of venv/bin/python.
+    if Path(prog).name == "env":
```
andrexibiza, line 120: **Blocker — this blanket `env` branch converts a valid launcher into a broken one.** `env` does not always mean PATH-dependent `env python3`: GNU `env -S` can pin an absolute venv interpreter and its required flags, e.g. `#!/usr/bin/env -S /opt/hermes/venv/bin/python -I`.

At this head, with `sys.executable == /usr/bin/python3`, the baseline classifier leaves that launcher direct, but this branch returns `True` solely because `prog` is `env` and emits `/usr/bin/python3 <script>`. In an isolated executable repro, the direct shebang exits 0 and imports a dependency installed only in the pinned venv; the emitted command exits 1 with `ModuleNotFoundError`. It also discards `-I` because Python is now handed the script as an argument.

Parse the original-case shebang tokens. Prefix only PATH-dependent `env python*` forms; for `env -S` with an absolute Python target, compare that target by path components/environment identity and preserve its flags. Add the clean-environment subprocess regression.

andrexibiza (commented):
Reviewed exact head `9d268ffe5224ee8176b82c450d4daa7dbed2c189` against `main@3f36c87e1ebdfbf7d14a88229dc9be222c12ea89`.

Keeping `sys.executable` lexical rather than resolving through the venv symlink is the correct repair for the reported uv escape, and the new symlink regression binds that positive case. One runtime blocker remains in the broadened shebang classification; I left it inline.

Fresh-state adversarial receipt, using real interpreters and a clean desktop-style environment:

```text
uv symlink positive control: candidate needs prefix = False
env -S absolute-venv shebang: baseline needs prefix = False
env -S absolute-venv shebang: candidate needs prefix = True
direct valid shebang under env -i: exit 0, stdout = venv-ok
candidate-emitted /usr/bin/python3 <script>: exit 1, ModuleNotFoundError
```

The parser must distinguish PATH-dependent `#!/usr/bin/env python...` from `env -S` forms that already pin an absolute interpreter, preserve path case, and retain interpreter flags. Add the executable regression, not only a rendered-string assertion.

Before merge, consolidate the existing implementation graph rather than landing another synonymous repair. Prior work by GitTradWang in #94058/#94051, jackulau in #92090, and gokhanyildirimlar in #94874 already covers parts of this exact parser and verification matrix; mojtabazn's #97983 is the new duplicate report. Preserve those credits and add the repo-template closing interlock to the selected canonical issue.

Exact-head CI is not green evidence yet: the `CI` run is `action_required` with zero jobs executed: https://github.com/NousResearch/hermes-agent/actions/runs/33261474514

PR #97999, opened 2026-08-29
fix(honcho): switching setup to API-key auth retires the OAuth grant

### LINKED ISSUE #97990: hermes honcho setup apikey mode doesn't clear stale OAuth block, auth silently stays broken

SilverNine:
Fixes #97990 — `hermes honcho setup` → `apikey` now actually takes effect after an OAuth grant is revoked.

The reporter's diagnosis was right, and the mechanism turns out to be a **write/read asymmetry on the host block** rather than anything in the wizard's prompt flow:

| | writes `hosts.<host>.apiKey` | writes `hosts.<host>.oauth` | writes top-level `apiKey` |
|---|---|---|---|
| OAuth path (`oauth._persist_credential`) | ✅ (the access token) | ✅ | — |
| API-key path (before this PR) | — | — | ✅ |

Every reader prefers the host block:

- `cli._resolve_api_key` reads `_host_block(cfg, _host_key()).get("apiKey")` **before** `cfg["apiKey"]`
- `client._credential_fingerprint` keys on `hosts.<host>.oauth.refreshToken` whenever that block is present, so the client cache identity is pinned to the grant

So the pasted key was written somewhere nothing reads first. The revoked access token kept winning the key lookup, the revoked refresh token kept deciding cache identity, and both survived every `honcho.json` rewrite because nothing on the API-key path cleared them — which is exactly why repeated completed setup runs changed nothing and only a manual edit worked.

### The fix

`_apply_api_key_to_host()` writes the key where the resolver actually reads it (host block **and** top level, the latter kept for legacy/global readers) and drops the superseded `oauth` block. Identity/workspace settings and other hosts are untouched — only this host's credential is replaced.

One deliberate scope decision worth review: it runs for a **kept** key too (blank prompt = keep current), not only for a newly pasted one. The shadowing survives config rewrites, so re-running setup has to be able to clear it; requiring the user to re-paste a key they already have in order to escape a dead grant would be a strange recovery path. That also repairs a smaller pre-existing case — a blank prompt with a valid top-level key previously left the host block untouched entirely.

I deliberately did **not** change the resolution order itself. Host-over-root precedence is load-bearing for profile isolation, and the defect is that the API-key path never wrote at that level — not that the precedence is wrong.

### Verification

Sabotage-checked rather than assumed. Reverting the helper to the previous top-level-only write reproduces the reported symptom exactly, in all three of its observable forms:

```
AssertionError: assert 'hon_oauth_access_REVOKED' == 'hon_sk_freshly_pasted_key'
AssertionError: the dead grant survived; status will keep reporting Auth: OAuth
AssertionError: identity still derives from the revoked refresh token; a cached
                OAuth client would be reused for the new key
4 failed, 2 passed
```

Restored: 6 passed. The six regressions assert behaviour contracts rather than the config's shape — that the key resolves through `_resolve_api_key`, that the credential fingerprint moves off the dead grant and lands on the pasted key's identity, that peer/workspace settings survive, and that an unrelated OAuth host keeps its grant.

`tests/honcho_plugin`: **311 passed, 16 skipped**. `tests -k honcho`: **375 passed, 20 skipped**.

### Not covered

I could not reproduce the original server-side revocation end to end — I have no revoked Honcho grant to hand, so the starting state is reconstructed from the on-disk shape the OAuth path writes (`apiKey` + `oauth` with a stale `expiresAt`) rather than captured from a live `invalid_grant`. What is proven here is that the config the reporter described resolves to the wrong credential before this change and the right one after. If a maintainer would rather see the wizard exercised end to end than the helper it delegates to, I am happy to add that.

### DIFF
--- a/plugins/memory/honcho/cli.py
+++ b/plugins/memory/honcho/cli.py
@@ -277,6 +277,30 @@ def _write_config(cfg: dict, path: Path | None = None) -> None:
     atomic_json_write(path, cfg, mode=0o600)
 
 
+def _apply_api_key_to_host(cfg: dict, host: str, api_key: str) -> None:
+    """Make ``api_key`` the effective credential for ``host``.
+
+    Choosing API-key auth has to retire the OAuth grant, not sit beside it
+    (#97990).  The OAuth path persists *two* things into the host block —
+    ``apiKey`` (the access token) and ``oauth`` — while readers prefer the
+    host block over the top level: ``_resolve_api_key`` reads
+    ``hosts.<host>.apiKey`` first, and ``client._credential_fingerprint``
+    keys on ``oauth.refreshToken`` whenever that block exists.  Writing only
+    the top-level ``apiKey`` therefore leaves a revoked grant winning both
+    lookups, and ``hermes honcho status`` keeps reporting ``Auth: OAuth``
+    across repeated, fully completed setup runs.
+
+    So write the key where the resolver actually reads it, keep the
+    top-level copy for the legacy/global readers, and drop the superseded
+    ``oauth`` block.  Identity/workspace settings and other hosts are left
+    untouched — only the credential for this host is replaced.
+    """
+    cfg["apiKey"] = api_key
+    block = cfg.setdefault("hosts", {}).setdefault(host, {})
+    block["apiKey"] = api_key
+    block.pop("oauth", None)
+
+
 def _resolve_api_key(cfg: dict) -> str:
     """Resolve API key with host -> root -> env fallback.
 
@@ -746,13 +770,19 @@ def cmd_setup(args) -> None:
             masked = f"...{current_key[-8:]}" if len(current_key) > 8 else ("set" if current_key else "not set")
             print(f"\n  Current API key: {masked}")
             new_key = _prompt("Honcho API key (leave blank to keep current)", secret=True)
-            if new_key:
-                cfg["apiKey"] = new_key
 
-            if not cfg.get("apiKey"):
+            effective_key = new_key or current_key
+            if not effective_key:
                 print("\n  No API key configured. Get yours at https://app.honcho.dev")
                 print("  Run 'hermes honcho setup' again once you have a key.\n")
                 return
+
+            # Choosing API-key auth retires any OAuth grant on this host: the
+            # host block outranks the top level in every reader, so a
+            # top-level-only write leaves a revoked grant deciding auth
+            # (#97990).  Done for a kept key too — the shadowing survives
+            # config rewrites, so re-running setup has to be able to clear it.
+            _apply_api_key_to_host(cfg, _host_key(), effective_key)
 
     # --- 3. Identity ---
     current_peer = hermes_host.get("peerName") or cfg.get("peerName", "")
--- a/tests/honcho_plugin/test_apikey_retires_revoked_oauth.py
+++ b/tests/honcho_plugin/test_apikey_retires_revoked_oauth.py
@@ -1,0 +1,178 @@
+"""Switching `hermes honcho setup` to API-key auth must retire the OAuth grant.
+
+Reported in #97990: after a Honcho OAuth grant is revoked server-side
+(`invalid_grant`), running `hermes honcho setup` -> `apikey` and pasting a
+valid key leaves authentication broken. `hermes honcho status` keeps
+reporting `Auth: OAuth (...)` across repeated, fully completed setup runs.
+
+Root cause is a write/read asymmetry in the host block:
+
+  * the OAuth path persists BOTH `hosts.<host>.apiKey` (the access token) and
+    `hosts.<host>.oauth` (`oauth._persist_credential`),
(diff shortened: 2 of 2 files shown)


Enough1122:

### PR #97999 — fix(honcho): switching setup to API-key auth retires the OAuth grant

**Summary:** Honcho setup's API-key branch now calls `_apply_api_key_to_host(cfg, host, effective_key)` which writes `apiKey` at both top-level and `hosts.<host>.apiKey` and drops `hosts.<host>.oauth`. This fixes #97990 where `_resolve_api_key` and `client._credential_fingerprint` prefer the host block, so a top-level-only write left a revoked OAuth grant winning across repeated setup runs (`Auth: OAuth` stuck). Also applies when keeping the current key (re-running setup clears lingering `oauth` shadow).

- `_apply_api_key_to_host` uses `setdefault("hosts", {})` / `setdefault(host, {})` — preserves other hosts and identity/workspace settings. `hosts.<host>.pop("oauth", None)` is safe idempotency.
- Retiring on "keep current" may drop an `oauth` the user intended to keep, but the explicit choice of the API-key branch signals intent to use API-key auth — correct to clear shadow.

**Verdict:** Correct credential retirement fix. No issues.

SilverNine:
Status check from my side, with two things that may be useful for triage rather than just a ping.

**1. Still applies cleanly to today's `main`.** I re-diffed the head of this branch (`4bac71f`) against `main` at `1802911` a moment ago:

```
plugins/memory/honcho/cli.py | 36 +++++++++++++++++++++++++++++++++---
1 file changed, 33 insertions(+), 3 deletions(-)
```

That is the same `+33/-3` as when the PR was opened, i.e. **zero drift** — the last change to `plugins/memory/honcho/cli.py` on `main` is `864035b` (Aug 7), well before this branch's base, so nothing has landed near the touched code. The other file is a new test that collides with nothing.

**2. The linked issue is still labelled `needs-repro` — this PR carries the repro.** `#97990` has no comments and the `needs-repro` label, but the reproduction it asks for is already in `tests/honcho_plugin/test_apikey_retires_revoked_oauth.py`, sabotage-checked: reverting the helper to the previous top-level-only write reproduces all three observable forms of the report (key resolves to the revoked access token, `Auth: OAuth` persists, credential fingerprint stays pinned to the dead refresh token), and restoring it turns all six green. If it helps, the label on `#97990` can probably come off on that basis regardless of what happens to this PR.

For the record on the one caveat I flagged in the description: I still have no live server-side-revoked grant, so the starting state is reconstructed from the on-disk shape the OAuth path writes, not captured from a real `invalid_grant`. What is proven is the resolution behaviour before and after, not the revocation event itself. If a maintainer would rather see the wizard driven end to end than the helper it delegates to, say the word and I will add it.

No review has been requested on this yet and there is no assigned reviewer — could someone take a look, or point me at the right person for `plugins/memory/honcho`? Happy to rebase or re-scope if the "kept key also retires the grant" decision is the part that gives pause; that is the one deliberate judgement call in here.

kshitijk4poor:
Fixed on main via #109326 (merge 135221e216), salvaged from #97364 which was opened a day earlier and carried the same fix: the wizard's api-key answer drops the host's `oauth` block (`cli.py` `hermes_host.pop("oauth", None)`) and `enabled: true` is written only when honcho.json itself resolves a credential. Your diagnosis of the write/read asymmetry was right — thank you @SilverNine. #97990 is closed against that merge.

PR #88450, opened 2026-08-17
i18n(nav): route the remaining sidebar entries through i18n

mikhailchuklay:
## What does this PR do?

Six sidebar entries — **Files, MCP, Channels, Webhooks, Pairing, System** — never reach i18n, so they stay English in every locale while the rest of the sidebar translates. On a fully translated dashboard this looks like a half-finished UI.

The mechanism for this already exists and is used by every other entry:

```ts
(t.app.nav as Record<string, string>)[labelKey] ?? label
```

Those six items simply never got a `labelKey`. This PR gives them one and adds the strings.

## Related Issue

No existing issue — the gap is visible on any non-English dashboard.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `web/src/App.tsx` — added `labelKey` to the six entries in `BUILTIN_NAV_REST`. `label` stays as the fallback, so behaviour is unchanged when a locale lacks the key.
- `web/src/i18n/types.ts` — added `channels`, `files`, `mcp`, `pairing`, `system`, `webhooks` to `app.nav`, **as optional fields**.
- `web/src/i18n/en.ts` — added the six English strings.
- `web/src/i18n/ru.ts` — added the six Russian strings.

**Why optional:** the lookup goes through `Record<string, string>` with a `?? label` fallback, so a missing key degrades to English rather than breaking. Making the fields optional keeps the other 15 locales valid and lets translators pick these up whenever convenient, instead of forcing 15 files into this PR. Happy to make them required and fill every locale if you'd prefer that.

`MCP` is left untranslated in Russian — it is a proper noun.

## How to Test

1. `cd web && npx tsc --noEmit -p tsconfig.json` — passes (verified against the Docker image build).
2. Open the dashboard, switch language to any non-English locale.
3. Sidebar: `Files`, `MCP`, `Channels`, `Webhooks`, `Pairing`, `System` previously stayed English regardless of locale. With this change they follow the locale when the key exists, and fall back to English when it doesn't.

### Code

- [x] I've read the Contributing Guide
- [x] My commit messages follow Conventional Commits (`i18n(nav): …`)
- [x] I searched for existing PRs to make sure this isn't a duplicate
- [x] My PR contains **only** changes related to this fix
- [ ] I've run `pytest tests/ -q` — N/A, no Python touched; verified with `tsc --noEmit`
- [ ] I've added tests — N/A, no test harness covers sidebar labels
[...]

PR #51352, opened 2026-06-23
fix(web): add i18n labelKey for Files + System nav items

aser0709:
## Problem

The **Files** nav item in the sidebar was hardcoded as an English literal (`label: "Files"`) instead of using \`labelKey\`. This means the label never translates in non-English locales — users see "Files" regardless of their language setting.

The same issue affected the **System** nav item (fixed separately, already in working tree).

## Fix

- Added \`labelKey: "files"\` to the \`/files\` nav entry in \`App.tsx\`
- Added \`labelKey: "system"\` to the \`/system\` nav entry in \`App.tsx\`
- Added \`files\` key to i18n types (optional, with fallback to English)
- Added \`files\` translation to \`en.ts\` and \`ru.ts\`
- Added \`system\` translation to all 13 locale files for consistency

## Files changed

- \`web/src/App.tsx\` — 2 nav items now use \`labelKey\`
- \`web/src/i18n/types.ts\` — optional \`files\` key added
- \`web/src/i18n/en.ts\` — \`files\` + \`system\` added
- \`web/src/i18n/ru.ts\` — \`files: "Файлы"\` + \`system: "Система"\` added
- All other locale files — \`system\` fallback added

## Before / After

**Before (Russian locale):** sidebar shows "Files" (untranslated)
**After:** sidebar shows "Файлы" (translated)

PR #98001, opened 2026-08-29
fix(web): localize Files/Channels/Webhooks/Pairing/System nav labels

xmhuangzhijun-hue:
## What & why

Five sidebar entries in the dashboard are declared without a `labelKey`:

```ts
{ path: "/files",    label: "Files",    icon: FolderOpen },
{ path: "/channels", label: "Channels", icon: Radio },
{ path: "/webhooks", label: "Webhooks", icon: Webhook },
{ path: "/pairing",  label: "Pairing",  icon: ShieldCheck },
{ path: "/system",   label: "System",   icon: Wrench },
```

`NavItem` resolves its label with

```ts
const navLabel = labelKey
  ? ((t.app.nav as Record<string, string>)[labelKey] ?? label)
  : label;
```

so without a `labelKey` these five always fall back to the hardcoded English
`label` — they stay in English in **every** one of the 17 supported locales,
while the rest of the sidebar translates correctly. The `app.nav` type only
declares 12 keys, so there was nothing to point at either.

This PR:

- adds `channels`, `files`, `pairing`, `system`, `webhooks` to `app.nav` in
  `Translations` (keeping the block's existing alphabetical ordering),
- fills those five keys in all 17 locale files,
- wires the five nav entries to them via `labelKey`.

`MCP` is deliberately left as-is — it reads as a proper noun in every locale.

No behavior change for English users; no runtime or API change.

## How to test

1. `cd web && npm install && npm run build`
2. Start the dashboard and open it.
3. Switch the UI language (sidebar language button) to any non-English locale —
   e.g. 简体中文, 日本語, Deutsch.
4. **Before:** Files / Channels / Webhooks / Pairing / System stay English while
   Chat, Sessions, Models, … are translated.
   **After:** all six translate (MCP intentionally unchanged).

Verified in the `web` workspace:

| check | result |
|---|---|
| `npm run typecheck` | pass |
| `npm run lint` | 26 warnings, **0 errors** — same count as before this change |
| `npm run build` | pass |
| `npm run test` | 36 files / 278 tests pass |

No Python is touched, so the Python suite is unaffected (`git diff --stat` is
entirely under `web/src/`).

## Platforms

Developed and verified on Windows 11 (Node 24, npm 11). The change is pure
TypeScript data + one prop per nav entry, with no platform-dependent code.

## Note on translations

`en`, `zh`, `zh-hant`, `ja` and `ko` I'm reasonably confident in. The remaining
twelve locales are best-effort for five common UI nouns — **review from native
speakers is very welcome**, and I'm happy to amend any of them in this PR.

`webhooks` is intentionally left as the untranslated technical term in several
locales (e.g. `Webhooks` in zh/de/es/fr) where that is the conventional form;
where a localized form is idiomatic it is used (`웹훅`, `Вебхуки`, `Webhookok`,
`Webhook'lar`).


xmhuangzhijun-hue:
Closing as a duplicate of #88450, which was opened on 2026-08-17 — twelve days
before this one — and covers the same six nav entries with a lower-coupling
approach (optional type fields + the existing `?? label` fallback, so it does
not force every locale file to change at once). #51352 also touches part of
the same area.

My mistake: I verified that the bug was still present in `origin/main` and
concluded it was unclaimed, without searching the open PR queue first. Checking
the code only tells you it hasn't been merged, not that nobody is working on it.

I've offered the fifteen additional locale translations from this branch over
on #88450 in case they're useful there. Nothing here needs review — sorry for
the noise.

PR #98002, opened 2026-08-29
fix(desktop): stop a mirrored composer effort pinning every new chat

james47kjv:
## What

A new desktop chat is created with an explicit `reasoning_effort` per-session override taken from the composer's sticky selection — **even when that selection was never a user pick**. It is often just a *mirror* of some profile's `agent.reasoning_effort`, and in app-global remote mode routinely a **different** profile than the one the chat is being created against.

Result: a profile configured `agent.reasoning_effort: high` runs its desktop chats at `medium`, and the configured value is never consulted.

## Why it happens

The composer seeds its effort from config in `useHermesConfig`, but only while **no session is active**:

```ts
const shouldSeedComposer =
  !activeSessionIdRef.current &&
  getComposerSelectionGeneration() === selectionGeneration &&
  (force || getCurrentModelSource() !== 'manual')
```

In app-global remote mode one backend serves every profile. With a chat open — the normal state — switching to another bot never reseeds the pill, so it keeps the effort of whichever profile the app was last scoped to. `desktopSessionCreateParams` then ships that stale mirror as an explicit override:

```ts
...(selection.effort ? { reasoning_effort: selection.effort } : {}),
```

and `session.create` honours it as a per-session override, short-circuiting the target profile's configured value.

The value is also persisted to `localStorage` (`hermes.desktop.composer.reasoning-effort`) and re-written from `session.info` heartbeats, so once one `medium` chat exists the wrong effort sticks across restarts and spreads to new chats on every profile.

Observed on a 60-profile remote deployment: profiles pinned to `gpt-5.6-sol` / `high` in `config.yaml` had every desktop-created session persisted with `"reasoning_config": {"effort": "medium"}`, while `cron`-created sessions on the same profiles were correctly `high`.

## Why this fix

The selection source already distinguishes the two cases — this change just honours it, exactly as #91482 does for model/provider:

- `'manual'` → a real user pick → send it as a per-session override (unchanged)
- `'default'` / unset → a mirror of profile config → omit it, let the backend resolve the target profile's value

```ts
const isManualSelection = getCurrentModelSource() === 'manual'

const selection = {
  effort: isManualSelection ? $currentReasoningEffort.get().trim() : '',
  fast: $currentFastMode.get(),
  model: $currentModel.get().trim(),
  provider: $currentProvider.get().trim()
}
```

A manual effort pick still overrides: `patchReasoning` calls `markComposerSelectionManual()`, so the user's choice sets the source before it is read here.

## Scope

Deliberately scoped to `reasoning_effort` so it **composes with #91482** (same gate, model/provider half) rather than duplicating its diff. If #91482 lands first this rebases to the one `effort:` line — happy to squash them if you'd prefer a single PR.

`fast` still rides unconditionally, because the gateway documents its *presence* as an explicit pin contract ("omitted means inherit the profile, true pins priority, and false pins normal") and an existing test asserts `fast: false` is sent explicitly. Worth a separate look, but it is a deliberate contract rather than a leak.

## Reproduction

On a multi-profile backend:

1. Profile **A** has `agent.reasoning_effort: medium`, profile **B** has `high`.
2. Open a chat on **A** (the composer pill now reads Medium).
3. With that chat still open, switch to **B** and start a **new** chat.

**Before:** the new chat on **B** runs at `medium`; `hermes config get reasoning` for that session reports `medium` while B's `config.yaml` says `high`.
**After:** the new chat runs at **B**'s configured `high`.

### Second commit: the pill must not advertise what it will not send

Withholding the mirror from `session.create` left the composer free to display a
level the new chat would not use — it renders `reasoningEffort || defaultEffort`,
and on a draft `reasoningEffort` is that same mirror. It could read "Med" and
then start the chat on the profile's High.

A draft that is not running a manual pick now falls through to
`$defaultReasoningEffort`, which `useHermesConfig` publishes from profile config
on every refresh regardless of whether the composer is reseeded ("Publish the
profile default regardless of whether the composer is reseeded below"). A live
session still shows its own effort; a manual pick still shows the pick. This
matches the existing `pinnedOverride` badge, which already treats
`modelSource === 'manual'` as "this will override".

## How to test

Automated coverage — 4 new cases in the `createBackendSessionForSend profile routing` suite:

| Case | Asserts |
|---|---|
| `'default'`-sourced selection | `reasoning_effort` omitted |
| unset source (first run / cleared) | `reasoning_effort` omitted |
| manual pick | still sent as an override |
| model independence | `model` / `provider` still sent while effort is withheld |

…plus 3 in `ModelPill reasoning-effort label`: a mirror shows the profile
default, a manual pick shows the pick, and a live session keeps showing its own.

Three of them fail against the previous code with `expected { … } to not have property "reasoning_effort"`.

One existing test (`freezes the visible selector state before profile readiness`) now sets the selection source explicitly — it drives the atoms directly rather than going through a composer pick, so it opts into `'manual'` to keep asserting override behavior. Same adjustment #91482 needed.

## Verification

Run against this branch on `3f36c87e1e`:

- `npx vitest run` (entire desktop suite) — **8689 passed / 816 files**, 6 skipped
- `npx tsc -p tsconfig.json --noEmit` — clean
- `npx eslint` on all changed files — clean

## Platforms

Reproduced against a Windows (x64) packaged Electron build talking to a Linux backend over a remote gateway. The change is platform-independent renderer logic — no file I/O, process, or terminal handling — so no cross-platform impact is expected.

### DIFF
--- a/apps/desktop/src/app/chat/composer/model-pill.test.tsx
+++ b/apps/desktop/src/app/chat/composer/model-pill.test.tsx
@@ -4,7 +4,14 @@ import { afterEach, describe, expect, it } from 'vitest'
 
 import type { ChatBarState } from '@/app/chat/composer/types'
 import { type SessionView, SessionViewProvider } from '@/app/chat/session-view'
-import { $activeSessionId, $currentModel, setCurrentModel, setCurrentModelSource } from '@/store/session'
+import {
+  $activeSessionId,
+  $currentModel,
+  setCurrentModel,
+  setCurrentModelSource,
+  setCurrentReasoningEffort,
+  setDefaultReasoningEffort
+} from '@/store/session'
 
 import { ModelPill } from './model-pill'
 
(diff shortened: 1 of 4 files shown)


Enough1122:

### PR #98002 — fix(desktop): stop a mirrored composer effort pinning every new chat

**Summary:** `ModelPill` previously displayed the draft's sticky `currentReasoningEffort` for every new chat, even though the draft only sends that effort when `modelSource==='manual'`. Now computes `pinnedOverride`/`view.kind` logic so a `primary` draft with no `runtimeId` and `source==='default'` shows `defaultReasoningEffort` (profile default) instead; manual picks and live sessions (`$activeSessionId !== null`) still show the sticky effort. Adds 3 tests: mirror → High, manual → Med, live session → Med.

- Uses `setDefaultReasoningEffort` store which syncs from gateway; if default is empty the pill falls back to sticky — safe.
- Fixes mirrored effort leaking across new chats as described (#?).

**Verdict:** Narrow UI fix with good coverage. No issues.

PR #98003, opened 2026-08-29
fix(guardrails): add crash-class hint to terminal failure recovery

Rmohid:
## Problem

When a terminal command fails with an import crash — `ModuleNotFoundError`, ABI mismatch from a compiled `.so` built for another CPython, PYTHONPATH-poisoned interpreter — the agent re-runs the same failing command repeatedly (observed: 46× in one session) because the failure reads as "need to search more" rather than "environment is broken".

Bead: worktracker-676.4.22.175.57

## Change

- `_tool_failure_recovery_hint` (terminal branch) now appends a crash-class hint: check the environment first; a PYTHONPATH-poisoned interpreter re-reports the same crash no matter how many times the command is re-run.
- `configuration.md` clarifies that `tool_loop_guardrails` counters reset at the start of every turn, so a legitimate multi-turn session is never starved.

## Verification

- `python3 -m pytest tests/agent/test_tool_guardrails.py -q` → 7 passed
- Manual check: `_tool_failure_recovery_hint('terminal', 4)` includes the crash hint; non-terminal tools unchanged.

### DIFF
--- a/agent/tool_guardrails.py
+++ b/agent/tool_guardrails.py
@@ -759,6 +759,10 @@ def _tool_failure_recovery_hint(tool_name: str, count: int) -> str:
             "For terminal failures, run a small diagnostic such as `pwd && ls -la` "
             "in the same tool, then try an absolute path, a simpler command, a different "
             "working directory, or a different tool such as read_file/write_file/patch."
+            " If the error is an import crash (ModuleNotFoundError, ABI mismatch, "
+            "a compiled .so built for another CPython), check the environment first: "
+            "a PYTHONPATH-poisoned interpreter re-reports the same crash no matter "
+            "how many times the command is re-run — fix the env, not the command."
         )
     return common + (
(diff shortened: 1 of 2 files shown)


Enough1122:

### PR #98003 — fix(guardrails): add crash-class hint to terminal failure recovery

**Summary:** Extends `_tool_failure_recovery_hint` terminal failure branch with: if `ModuleNotFoundError`/ABI/`.so` built for another CPython, check env first — `PYTHONPATH`-poisoned interpreter re-reports same crash, fix env not command. Docs note in `configuration.md` that counters reset each turn.

- String appended without trailing period duplication — check spacing: existing hint ends without trailing newline, added text starts with leading space via `"` If..."` — correct.
- Docs change clarifies `hard_stop_enabled` per-turn semantics — helpful.

**Verdict:** Docs/hint improvement. No issues.

PR #98009, opened 2026-08-29
fix(cli): apply --in DIR for one-shot (-z) runs

barnscarter-ops:
`--in` parses to `args.in_dir`, but its only consumer is `cmd_chat`, which chdirs into the directory. All three `-z`/`--oneshot` dispatch sites call `_run_and_exit_oneshot()` and exit without ever reaching `cmd_chat`, so the flag is parsed, validated, and then silently discarded.

With no chdir and no registered workspace override, relative paths in the file tools fall through to their last anchor, `os.getcwd()`. A one-shot asked to create a file at a relative path writes it into the launcher's working directory instead of the assigned workspace — while still reporting success.

## Repro

```
hermes -z "Create a file named SUMMARY.md here" --in /tmp/workspace
```

replies `DONE`, and `SUMMARY.md` lands in the launcher's cwd.

## Fix

Lift the existing chdir out of `cmd_chat` into a shared `_enter_in_dir()` helper, call it from `_run_and_exit_oneshot` before the agent is constructed, and forward `in_dir` from all three dispatch sites.

`cmd_chat` behaviour is unchanged: same messages, same stdout stream, same `no_restore_cwd` side effect. The one-shot path sends its error to stderr because `-z` reserves stdout for the final response.

I considered registering a workspace override into the file tools instead of chdir'ing, but the CLI one-shot is always a dedicated, short-lived process (`_run_and_exit_oneshot` ends in `os._exit()` in a `finally`), so a process-global chdir cannot leak into another session. It also matches the established `--in` semantics for `chat`, and fixes reads, edits, globs and the terminal tool in one move.

## Tests

Four added, including a source-level guard that fails if a new `-z` dispatch site is added without forwarding `in_dir`.

Measured on `tests/hermes_cli/test_resume_latest_and_in_dir.py` at this branch's merge base:

| | result |
|---|---|
| without the fix | 4 failed, 12 passed |
| with the fix | 1 failed, 15 passed |

The remaining failure, `test_in_dir_expands_user_home`, is **pre-existing and unrelated** — it monkeypatches `HOME`, but `os.path.expanduser` on native Windows reads `USERPROFILE`. It fails identically with and without this change, and I left it alone rather than widen the scope.

## Behaviour changes for external `-z --in` callers

- A bogus `--in` now fails the run (exit 1, stderr) instead of silently running in the wrong directory.
- A *relative* `--usage-file` now resolves inside the `--in` workspace.

No caller inside the repo is affected: every internal `--in` use goes through the `chat` subcommand, which already chdirs and is unchanged.

### DIFF
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -174,6 +174,45 @@ def _cleanup_oneshot_runtime() -> None:
         pass
 
 
+def _enter_in_dir(in_dir: object, *, stream=None) -> str | None:
+    """Apply ``--in DIR``: chdir into DIR, or exit(1) with a clear message.
+
+    Shared by ``cmd_chat`` and ``_run_and_exit_oneshot`` so ``--in`` means the
+    same thing for every launch mode. The process cwd is the LAST anchor the
+    file tools fall back to when resolving a relative path
+    (``tools.file_tools._resolve_base_dir``, step 4), so a launch mode that
+    parsed ``--in`` but never chdir'd wrote relative paths into the launcher's
+    cwd instead of the assigned workspace.
+
+    ``stream`` selects where the error text goes; defaults to stdout to match
+    the rest of the CLI. ``-z`` passes stderr because one-shot reserves stdout
+    for the final response.
+
+    Returns the absolute directory entered, or ``None`` when *in_dir* is unset.
+    """
+    if not in_dir:
+        return None
+    # Git Bash / MSYS hands the CLI POSIX-style paths (`--in ~` expands to
+    # `/c/Users/x` before Python ever sees it; MSYS2's path conversion is
+    # disabled for native executables). Translate the MSYS/Cygwin/WSL
+    # drive-root spellings to native Windows form first — no-op elsewhere.
+    from tools.environments.local import _msys_to_windows_path
+
+    _target_dir = os.path.abspath(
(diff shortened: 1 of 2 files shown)


Enough1122:

### PR #98009 — fix(cli): apply --in DIR for one-shot (-z) runs

**Summary:** Shares `_enter_in_dir(in_dir, stream=None)` between `cmd_chat` and `_run_and_exit_oneshot`. One-shot (`-z`) previously parsed `--in` but never `chdir`'d, so relative file writes landed in launcher cwd. Now `chdir`s before agent construction so session cwd record and `tools.file_tools._resolve_base_dir` (fallback step 4) both anchor correctly. Handles MSYS/Cygwin/WSL drive roots via `_msys_to_windows_path`, expands `~`, aborts with clear message on missing dir or `chdir` `OSError`. Errors go to `stderr` for `-z` (reserves stdout for final response).

- `os.chdir` is process-wide; subsequent code runs in target dir as intended for both chat and one-shot. `in_dir`/`_target_dir` are scoped locals — no global pollution.
- `_msys_to_windows_path` is imported inside function to avoid top-level cycle — fine.
- `_enter_in_dir` called before `run_oneshot` import so failure exits before session creation.

**Verdict:** Correct flag-application fix for `--in` on `-z`. No blocking issues.

PR #98017, opened 2026-08-29
fix(state): quarantine 0-byte truncated state.db (salvage of #97580)

### LINKED ISSUE #97568: is_zeroed_state_db() returns False for a 0-byte file, so total data loss is silent
### Summary

`is_zeroed_state_db()` returns `False` for a file of size `0`, so the zeroed-file quarantine path
never fires for the one case where the loss is total. A `state.db` truncated to zero bytes is
handled identically to a brand-new install: no ERROR log, no `state.db.quarantine-*` rename, no
signal of any kind.

```python
# hermes_state.py, is_zeroed_state_db()
if size <= 0:
    return False
```

The name of the function says what it is for. The guard excludes the most complete form of the
condition it is named after.

### Why it matters more than it looks

The quarantine path is good: it logs at ERROR and renames the bad file aside before opening a fresh
one, which is exactly the loud, forensic behaviour an operator needs. Everything downstream of that
`return False` is therefore skipped precisely when the damage is greatest.

And the resulting state is silent by construction. A store that lost every message and a store that
was created five minutes ago are byte-for-byte indistinguishable afterwards: same empty schema, same
`schema_version`, a virgin `sqlite_sequence`, ids restarting at 1. Nothing an operator can query
tells them apart.

Observed on 0.19.1 (`v2026.7.30`): a deployment lost its entire message history and it was not
noticed for **seven days**, because there was nothing to notice. The database file's own birth time
was a full three weeks earlier than its oldest surviving message — the only surviving hint, and one
that does not survive a copy or a restore.

### Suggested fix

Two independent parts, both small:

1. Treat `size == 0` as zeroed. A file that exists and holds nothing is the clearest possible case;
   only a genuinely absent file should take the "new install" path. If the current behaviour is
   deliberate — because a first run legitimately creates a 0-byte file before the schema is written —
   then the discriminator needs to be something other than size, and the comment should say so,
   because as written the guard reads like an oversight.

2. Record provenance at creation. A row in `state_meta` written in the same transaction as
   `schema_version` — a `store_instance_id` and a `store_created_at_utc` — would make "young" and
   "emptied" distinguishable forever, at the cost of two inserts. There is currently no hook,
   plugin or extension point between opening the database and `_init_schema()`, so this specific
   fix can only come from upstream: `hermes_state.py` opens SQLite and calls `self._init_schema()`
   immediately, and `hermes_state_schema.py` runs the schema and sets `schema_version` with no
   callback in between.

The second is the more valuable of the two, and it is also the one nobody downstream can implement.

kshitijk4poor:
## Summary

Salvage of PR #97580 by @loulanyue onto current `main`, authorship preserved. Fixes #97568.

`is_zeroed_state_db()` returned False for a 0-byte file (`size <= 0` guard), so a fully truncated state.db skipped quarantine and SQLite silently re-initialized an empty schema over the loss. 0-byte files are now classified as zeroed and quarantined with provenance recorded.

## Validation
| | result |
|---|---|
| tests/test_zeroed_state_db.py | 5 passed |
| mutation (hermes_state.py + siblings reverted to main) | 1 failed → restored green |

Closes #97580
Fixes #97568

### DIFF
--- a/hermes_cli/backup.py
+++ b/hermes_cli/backup.py
@@ -449,15 +449,17 @@ def is_zeroed_sqlite_file(
         size = path.stat().st_size
     except OSError:
         return False
-    if size <= 0:
+    if size < 0:
         return False
     from hermes_cli.sqlite_safe_read import read_header_bytes_preopen
 
     head = read_header_bytes_preopen(
         path, length=max(16, probe_bytes), force=force
     )
-    if not head:
+    if head is None:
         return False
+    if len(head) == 0:
+        return True
     if head.startswith(b"SQLite format 3"):
         return False
     return all(byte == 0 for byte in head)
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -3952,14 +3952,18 @@ def is_zeroed_state_db(
         size = path.stat().st_size
     except OSError:
         return False
-    if size <= 0:
+    if size < 0:
         return False
     from hermes_cli.sqlite_safe_read import read_header_bytes_preopen
 
     head = read_header_bytes_preopen(
         path, length=max(16, probe_bytes), force=force
     )
-    if not head or head.startswith(b"SQLite format 3"):
+    if head is None:
+        return False
+    if len(head) == 0:
+        return True
+    if head.startswith(b"SQLite format 3"):
         return False
     return all(byte == 0 for byte in head)
 
--- a/hermes_state_schema.py
+++ b/hermes_state_schema.py
@@ -8,11 +8,14 @@ own; methods access the host's attributes (``self._conn``, ``self.db_path``,
 module-level constants live in hermes_state_common.
 """
 
+import datetime
 import logging
 import json
 import sqlite3
 import time
+import uuid
 from typing import Dict, Optional, Sequence
+
 
 from hermes_constants import get_hermes_home
 from hermes_state_common import (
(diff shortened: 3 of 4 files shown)


egilewski:
suggesting changes

The patch correctly recognizes and quarantines a zero-byte regular state.db, but it can rename an inode already opened by another process and it blocks on zero-sized special files. The resulting write failure, divergent replacement database, and startup hang require changes.

- [P1] Zero-byte quarantine races with another process's open database (`hermes_state.py:3955`)
  SessionDB probes and renames a zero-byte path under only a process-local connection registry and a quarantine lock. A second process can already hold an SQLite connection to that inode; the rename then makes its journal path unavailable, causing a write error while the replacement state.db lacks the in-flight state. Session state can split or be lost.
  Remediation: Coordinate quarantine with an inter-process database/open lock or prove no other process has opened the inode before rename; never replace a path that may have a live SQLite connection.

- [P2] Zero-byte probe blocks on named pipes (`hermes_cli/backup.py:452`)
  Changing the size guard from <= 0 to < 0 sends zero-sized special files through read_header_bytes_preopen. Opening a FIFO for reading blocks until a writer arrives, so a local process that replaces state.db with a FIFO can hang SessionDB startup indefinitely.
  Remediation: Require a regular file before opening it and reject special files without performing a blocking read.

Security evidence:
- trust boundary: An untrusted local process or concurrent Hermes component can control or open the state.db pathname before schema initialization; SessionDB then probes bytes, quarantines, and recreates that path.
- source/sink/invariant: A state database must not be renamed or replaced while another process may write the opened inode, and a pre-open byte probe must reject non-regular files without blocking.
- current-main reproduction: On current main, an empty regular file is not classified as zeroed, the concurrent opener writes through the shared database, and the FIFO helper returns False while SessionDB fails fast with a disk I/O error.
- PR-head or patch-replay validation: On the patch replay, the zero-byte regular-file tests pass, but a second process holding an empty-file connection receives attempt to write a readonly database after quarantine and the live replacement has no marker; a FIFO helper and SessionDB probe both time out.
- positive/negative cases: Positive: the patch targeted suite quarantines a zero-byte regular file and records provenance. Negative controls: valid SQLite remains non-zeroed, current-main concurrent startup preserves its marker, and current-main FIFO handling returns promptly.
- residual bypass search: Both duplicate zeroed-file helpers, the SessionDB preflight/quarantine caller, concurrent startup, special-file handling, and schema metadata insertion were reviewed; no other changed sink was found.
- reviewer validation: Offline bounded probes used the archived current-main tree and staged patch replay; targeted tests passed on both, while the race and special-file probes failed only on the patch replay.

Not checked:
- CodeRabbit
- full test suite

PR #98018, opened 2026-08-29
fix(profiles): make_targz writes temp-then-rename (salvage of #97835)

kshitijk4poor:
## Summary

Salvage of PR #97835 by @nftpoetrist onto current `main`, authorship preserved.

`make_targz()` wrote the archive at its final destination directly — a crash/full-disk mid-write left a corrupt `.tar.gz` at the path consumers trust (profile export/backup). It now writes to a sibling temp file and atomically renames on success; failures clean up the temp and leave any pre-existing archive untouched.

## Validation
| | result |
|---|---|
| tests/hermes_cli/test_archive_safe.py + test_profiles*.py | 78 passed, 2 skipped |
| mutation (archive_safe.py reverted to main) | 1 failed → restored green |

Closes #97835

### DIFF
--- a/hermes_cli/archive_safe.py
+++ b/hermes_cli/archive_safe.py
@@ -21,6 +21,7 @@ from __future__ import annotations
 import os
 import shutil
 import tarfile
+import tempfile
 from pathlib import Path, PurePosixPath, PureWindowsPath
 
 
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Focused salvage of #97835 — makes `make_targz` failure-atomic.

**What it does**
- `hermes_cli/archive_safe.py:16` — writes to `tempfile.mkstemp(dir=dest_dir, prefix=".archive_")` then `os.replace` on success; `BaseException` cleanup unlinks temp (`hermes_cli/archive_safe.py:39`).
- Correctly truncates only the temp, so a pre-existing `<base>.tar.gz` survives mid-`tf.add` failures (disk full / interrupt).
- Tests (`tests/hermes_cli/test_archive_safe.py:84`) pin preserve-on-failure, round-trip, and overwrite-on-success.

**Non-blocking notes**
- `os.fdopen(fd, "wb")` — if that wrapper itself raised, `fd` would leak (no `os.close(fd)` in except). In practice unlikely, but an `except` that also tries `os.close(fd)` would be tighter.
- Uses `os.path.dirname(archive_path) or "."` for sibling dir — good for bare basename case. `os.replace` is atomic on POSIX and Windows when on same filesystem (mkstemp in same dir guarantees it).

No functional issues found. Well-tested.

kvnloo (commented):
This one's already landed. The temp-then-rename fix in `make_targz` shipped to `main` via the original #97835, merged by teknium1 on 2026-08-31 — two days after this salvage was opened. The mechanism is unchanged in the salvage: same `mkstemp` sibling temp, same `os.replace` commit, same `BaseException` cleanup. The tests are unchanged too: `test_archive_safe.py` already exists on `main` with the same three cases (preserve-on-mid-write-failure, round-trip, overwrite-on-success). Recommending this be closed as superseded rather than merged.

Verified against current `main`. `make_targz` there already writes to `.archive_*.tar.gz.tmp` in the destination dir and renames onto `<base>.tar.gz` only after the archive is fully written, with the temp unlinked on any failure. Against `main`, the salvage's residual diff is purely cosmetic: a longer docstring, nested `with` blocks instead of `main`'s single multi-context `with`, and a try/except-`OSError` where `main` already has the tighter `with suppress(OSError)`. Merging would be docstring-and-whitespace churn that reformats `main`'s cleaner forms backwards.

**Note:** the body's "Closes #97835" keyword won't fire — #97835 is already closed (merged). The duplicate happened because the salvage was opened on 2026-08-29, while the original was still open, and the original then landed on 2026-08-31. Worth a habit check on future salvages: re-diff against `main` right before posting, since a merge in between silently empties the PR.

PR #96886, opened 2026-08-28
fix(state): classify the transient NULL SystemError as lock contention

Agi-Asi:
## What

A session transcript write can die with `SystemError('<connection> returned NULL without setting an exception')` — a CPython sqlite3 driver glitch under WAL after very long turns (reported after a 163-tool-turn / ~273k-token / 597-API-call session). The database itself is healthy: `PRAGMA integrity_check` passes, `BEGIN IMMEDIATE` + `ROLLBACK` succeed, disk has space. It is **not** `SQLITE_BUSY`, so the driver's busy-retry never sees it; the error rode up unclassified (`"unknown"`) and failed the turn with generic advice.

`classify_persistence_error` now maps the phrase to the **`locked`** bucket, so the turn explanation carries the same transient "session storage was busy — please send it again" guidance as lock contention. The next send re-flushes the unmarked tail through the existing all-or-nothing batch recovery contract, so no rows are lost.

## How to test

`pytest tests/run_agent/test_turn_completion_explainer.py -q` — 23 passed, including a new test asserting the exact reported `SystemError` text and a wrapped variant classify as `locked`, with an explicit no-collision check.

## Platforms

Linux (pytest). Classification + guidance change only; no storage-layer behaviour altered.

Fixes #94258

PR #98020, opened 2026-08-29
fix(state): classify the transient NULL SystemError as lock contention (salvage of #96886)

### LINKED ISSUE #94258: fix(state): retry transient NULL-without-exception SQLite SystemError
## Bug

A session transcript write can fail with a Python `SystemError` that is not handled by the session persistence retry path. The failure ends the turn with `session_persistence_failed` even though the state database is healthy and writable.

## Environment

- Hermes Agent: v0.20.5 (2026.8.19)
- Upstream commit: `ab0d9841`
- Model/provider during reproduction: `gpt-5.6-luna` via `openai-codex`
- OS: Linux 6.8.0-124-generic
- Database journal mode: WAL

## Observed error

```text
Session DB append_message failed: <hermes_cli.sqlite_safe_read.TrackedConnection object at 0x...> returned NULL without setting an exception

Turn ended with pending tool result (agent may appear stuck).
Turn ended: reason=session_persistence_failed
```

## Context at failure

The failure occurred at the end of an unusually long turn/session:

- 163 tool turns
- approximately 273k tokens of context
- 597 API calls in the session

The same `state.db` passed `PRAGMA integrity_check`, accepted `BEGIN IMMEDIATE` followed by `ROLLBACK`, and had available disk space.

## Expected behavior

A transient SQLite connection/driver failure should be retried using the same bounded retry policy as `database is locked`, `busy`, and `no more rows available` errors. If persistence ultimately fails, the user-facing diagnostic should identify the actual cause without losing the turn unnecessarily.

## Actual behavior

The exception is not an `sqlite3.Error` handled by `_execute_write()`. It propagates to `run_agent.py`, which logs `Session DB append_message failed` and ends the turn with `session_persistence_failed`.

## Suggested fix

Handle the Python `SystemError` message `returned NULL without setting an exception` in `hermes_state._execute_write()` as a bounded transient retry, after the transaction rollback. The retry should remain message-scoped and should re-raise unrelated `SystemError` instances.

A local mitigation was tested successfully with this narrow retry path. No database repair was required.

kshitijk4poor:
## Summary

Salvage of PR #96886 by @Agi-Asi onto current `main`, authorship preserved, plus the required AUTHOR_MAP entry.

The CPython sqlite3 glitch `SystemError: <connection> returned NULL without setting an exception` (seen after very long turns, e.g. 163-tool turns) fell through `classify_persistence_error()` to the generic unknown branch, so users got an unhelpful failure explainer. It's now classified as lock contention ("storage was busy — send it again"), matching the #94258 tail-recovery contract.

Scope note: this is deliberately the *classifier* fix only — the root cause (4 unlocked readers on the shared writer connection) is tracked separately via #90734, which we're staging with its exactly-once follow-up. This PR is symptom relief, not the fix, and does not close #94258/#85079.

## Validation
| | result |
|---|---|
| tests/run_agent/test_turn_completion_explainer.py | 23 passed |
| mutation (hermes_state.py reverted to main) | 1 failed → restored green |

Closes #96886


Enough1122:

**Overall:** Small, well-placed classification fix for the CPython/WAL `SystemError: returned NULL without setting an exception` glitch.

**What it does**
- `hermes_state.py:26` adds `if "returned null without setting an exception" in text: return "locked"` before the structural-corruption/disk buckets — correctly bucketed as transient contention with "storage was busy — send it again" guidance (#94258 tail-recovery).
- Preserves `unknown` for unrelated strings (`tests/run_agent/test_turn_completion_explainer.py:71`).

**Non-blocking notes**
- Lower-cased `text` match is intentionally broad — covers both `SystemError("<connection> returned NULL...")` and `sqlite3.OperationalError("statement returned NULL...")` with one phrase; won't collide with other buckets (placed after compression checks, before "database disk image is malformed").
- No new imports side-effects; string-only classifier stays pure.

No issues found.

teknium1:
Heads-up: the root-cause class fix for the unsynchronized reads landed for review in https://github.com/NousResearch/hermes-agent/pull/99502, and #96886 (the PR this salvages) was closed in its favor — with the reads synchronized, the bare `SystemError` no longer occurs on the persistence path, so the classifier mapping becomes dead code for that spelling. Leaving this open for a maintainer call on whether the classifier entry is still wanted as defense-in-depth.

kvnloo (commented):
A very long turn — 163 tool calls — could end with a cryptic failure instead of "storage was busy, send it again." The CPython sqlite3 glitch `SystemError: <connection> returned NULL without setting an exception` fell through to the generic unknown bucket, so the turn explainer gave repair-path advice for a database that was actually healthy. This files it under transient contention. Approving — and on the maintainer question in the thread: keep the entry as defense-in-depth.

The case for keeping it is misclassification risk, which is essentially zero. The phrase "returned NULL without setting an exception" is CPython-internals wording — no other failure can produce it, so this branch can never steal a genuinely corrupt or full-disk error the way a loose "locked" substring could. And the "locked" bucket's downstream semantics are exactly right for a transient driver glitch: the turn explainer (`agent/turn_explainers.py`) says "your message should already be saved — please send it again in a moment" instead of pointing at `hermes doctor` repair for a healthy database.

**Non-blocking:** the body is stale on the salvage premise. "Closes #96886" is a no-op — it's already closed in favor of #99502 — and the scope note points the root cause at #90734, but the root-cause class fix already landed as #99502. The honest framing for this PR is defense-in-depth against a driver-level glitch the lock fix can't rule out; the body should say that.

**Non-blocking:** I checked the other consumer of this bucket. `hermes_state_compression.py:565` treats "locked" as "keep polling within the wait budget" in the turn-lease wait loop, so the new branch also re-routes the `sqlite3.OperationalError` spelling there — previously re-raised immediately, now waits and retries. That matches the transient semantics; not a blocker, but the author should know the change isn't guidance-only. (The `SystemError` spelling never reaches that loop — it's not an `sqlite3.Error` — so for it the change is guidance-only.)

**Non-blocking:** the code comment still frames the trigger as live ("surfaces under WAL after very long turns"). With the unsynchronized-read fix merged, one line saying this entry is defense-in-depth would keep future readers from re-chasing the trigger.

One pattern thread, offered not required: the classifier buckets do double duty. The docstring frames them as guidance selection, but "locked" also buys up-to-30-minutes patience in the lease waiter — a classification change is silently a retry-policy change. Worth one line in the `classify_persistence_error` docstring naming which buckets carry control-flow meaning, so the next classifier tweak doesn't surprise anyone.

PR #96503, opened 2026-08-27
fix(packaging): include wheel in PEP 517 build-system requires

olympusbuildz:
## Summary
Windows PowerShell installer editable installs fail in uv's PEP 517 build-isolation sandbox with `ModuleNotFoundError: No module named 'wheel.cli'` (and a follow-on `pkg_resources` miss) even when the real venv already has `wheel` installed.

## Root cause
`[build-system].requires` listed only `setuptools==83.0.0`. uv seeds the isolation environment from that list alone. `setuptools.build_meta` and our `setup.py` bdist_wheel guard import `wheel` during the build, so the sandbox cannot import `wheel.cli`.

## Fix
- Add `wheel` to `[build-system].requires`
- Add `wheel = false` to `[tool.uv].exclude-newer-package` so the existing build-system exclude-newer brick guard (`test_build_system_requires_exempt_from_exclude_newer`) remains green
- Regression test: `test_build_system_requires_wheel_for_isolated_builds`

## Why it matters
Blocks fresh Windows installs via `install.ps1` / `uv sync --extra all --locked` before the agent can launch. Workaround `UV_NO_BUILD_ISOLATION=true` works but should not be required.

## Test plan
```bash
python -m pytest tests/test_packaging_metadata.py -q
# 8 passed @ 715c96ba83
```

## Risk / blast radius
Build-metadata only. Does not change runtime deps. Unpinned `wheel` is the packaging toolchain; exclude-newer exemption matches the setuptools pattern for build requires.

## Closest work
none found (open PRs mentioning "wheel" are desktop scroll / unrelated)

Fixes #96488

Issue #96488, opened 2026-08-27
Windows installer fails at 'Installing Python dependencies' - ModuleNotFoundError: No module named 'wheel.cli' (build-system.requires missing wheel)


PCstepsGR:
**Environment**
- OS: Windows 11
- Python: 3.11.15 (uv-managed venv)
- uv: 0.12.6
- Hermes Agent: 0.20.6
- Installer: `scripts/install.ps1` (admin PowerShell), via `iex (irm https://hermes-agent.nousresearch.com/install.ps1)`

**Symptom**
The PowerShell installer aborts during the "Installing Python dependencies" stage, before node dependencies or the desktop build. The editable install (`uv sync --extra all --locked` and the `uv pip install -e .` fallback tiers) fails with:

```
ModuleNotFoundError: No module named 'wheel.cli'
ModuleNotFoundError: No module named 'pkg_resources'
```

The agent then fails to launch on next run because the editable install never completes.

**Root cause**
`pyproject.toml` declares:

```toml
[build-system]
requires = ["setuptools==83.0.0"]
build-backend = "setuptools.build_meta"
```

`wheel` is **not** in `build-system.requires`. uv builds the project backend in a PEP 517 build-isolation environment seeded only from that list, so the isolated sandbox contains `setuptools` but no `wheel`. The declared backend `setuptools.build_meta` imports `wheel.cli` during the editable build, and the repo's `setup.py` does `from setuptools.command.bdist_wheel import bdist_wheel` (which requires `wheel`). With `wheel` absent, the build fails with the two ModuleNotFoundErrors above.

The user's **real venv is fine** — `uv venv` seeds `pip`/`setuptools`/`wheel` into every venv it creates, and `wheel.cli`/`pkg_resources` import correctly there. The failure is specific to uv's build-isolation sandbox lacking `wheel`.

**Workaround (no source edit)**
Set `UV_NO_BUILD_ISOLATION=true` so uv builds the editable install using the already-seeded `wheel`/`setuptools` in the venv instead of the bare isolated sandbox:

```powershell
$env:UV_NO_BUILD_ISOLATION="true"; iex (irm https://hermes-agent.nousresearch.com/install.ps1)
```

**Proposed fix**
Add `wheel` to `build-system.requires` so the isolation environment has the packaging toolchain the build backend needs:

```toml
[build-system]
requires = ["setuptools==83.0.0", "wheel"]
build-backend = "setuptools.build_meta"
```

**Verification**
[...]

PR #98021, opened 2026-08-29
fix(packaging): wheel in PEP 517 requires + every exact pin exempt from exclude-newer (salvage of #96503 + #97955)

kshitijk4poor:
## Summary

Sequential salvage of the packaging pair — PR #96503 by @olympusbuildz and PR #97955 by @walker83 — onto current `main`, both authors' commits preserved. The two PRs edit the same `exclude-newer-package` region and could not merge independently.

1. **#96503**: `[build-system].requires` listed only setuptools, so uv's PEP 517 isolation sandbox lacked `wheel` — Windows installer editable builds failed with `ModuleNotFoundError: No module named 'wheel.cli'` even with wheel in the real venv (#96488). Adds `wheel` to requires + exclude-newer exemption + regression test.
2. **#97955**: every release exact-pins deps published days earlier; the relative `exclude-newer = "14 days"` cutoff filters those pins for ~2 weeks after release, bricking `hermes update` for any pre-release venv ("no version of snowballstemmer==3.1.1"). Converts the exemption dict to the table form exempting **every exact pin** (98 entries), enforced by a new test so future pins can't re-introduce the brick.

Conflict resolutions (both verified by TOML parse + tests): #96503 rebased over main's new `firecrawl-anydoc` entry (union); #97955's table gained the `wheel = false` entry from #96503.

## Validation
| | result |
|---|---|
| tests/test_packaging_metadata.py | 9 passed |
| mutation (pyproject.toml reverted to main) | 2 failed → restored green |
| TOML parse | 98 exclusion entries, wheel + firecrawl-anydoc both present |

Closes #96503
Closes #97955


egilewski:
suggesting changes

Two security gaps remain: the replay changes uv's exclude-newer package policy without updating uv.lock, so locked installs can fall into the unverified fallback, and it adds an unpinned wheel build requirement exempt from the age cutoff. Packaging guard tests pass but do not cover either invariant.

- [P1] Unpinned wheel bypasses the build freshness boundary (`pyproject.toml:388`)
  The replay adds bare wheel to build-system.requires and sets wheel=false under exclude-newer-package. setup.py imports setuptools.command.bdist_wheel during the PEP 517 build path, so an arbitrary newest wheel release can execute in every isolated build; the added test checks presence only, not a pin or lock hash.
  Remediation: Pin wheel to a reviewed version and retain its integrity metadata; exempt only exact-pinned build requirements from exclude-newer.

- [P1] Project resolver policy is not recorded in uv.lock (`pyproject.toml:459`)
  The replay expands tool.uv.exclude-newer-package from 13 to 98 entries but leaves uv.lock's options map unchanged. The installer tries uv sync --locked as its hash-verified tier and explicitly falls back to unverified PyPI resolution when that check fails, so the policy drift can remove the lockfile hash boundary for installs and updates.
  Remediation: Regenerate uv.lock from this pyproject so its resolver options match, and enforce uv lock --check in CI before using the locked install tier.

Security evidence:
- trust boundary: The trust boundary is the installer and PEP 517 build-isolation resolver: package metadata and artifacts come from an index, then build hooks execute before Hermes is installed. An index or release compromise can therefore run code in the build process.
- source/sink/invariant: The required invariant is that locked installation policy and project policy match, and every build-time package is pinned or remains constrained by the freshness cutoff; no floating build tool should bypass both controls.
- current-main reproduction: Current main has only setuptools==83.0.0 in build-system.requires and its canonical exclude-newer exception map matches the lock's 13 entries. The new wheel requirement and expanded map are absent.
- PR-head or patch-replay validation: Patch replay adds wheel without a version operator, records wheel=false, and expands the project exception map to 98 entries while the lock still records 13. The metadata guard suite passes, so it does not detect these gaps.
- positive/negative cases: The baseline snapshot passed seven packaging metadata tests and the replay passed nine; the replay lock-parity probe failed and the wheel exactness probe failed, while the baseline controls passed.
- residual bypass search: The exception table also contains floating-looking maturin and setuptools-rust entries; their concrete resolution was not exercised offline. No wheel package entry is present in uv.lock, so its artifact hash is not represented there.
- reviewer validation: Review used the bound current-main snapshot and staged patch replay, the bound Python executable, offline static probes, and pytest with caches and temporary files redirected to scratch.

Not checked:
- uv lock --check
- PEP 517 isolated build
- PEP 517 build guard
- network package resolution

teknium1:
Superseded by salvage PR #99493 — your consolidation was verified and is carried forward intact: all three commits (including your conflict resolutions between #96503 and #97955 and the attribution mapping) are cherry-picked with authorship preserved, and the mutation check from your validation table reproduces (revert pyproject → exactly the 2 new guards fail).

One gap required a new base: this branch changes the `exclude-newer-package` policy without regenerating `uv.lock`, and `uv lock --check` fails on it ("Resolving despite existing lockfile due to addition of exclude newer exclusion for package `elevenlabs`") — meaning the Tier-0 hash-verified `uv sync --locked` install path would break on merge. #99493 adds the regenerated lock (verified metadata-only: options table + marker refinements, zero resolved version or sha256 changes). Thanks @kshitijk4poor — the consolidation and sequential-cherry-pick attribution work made the final salvage trivial.

PR #98027, opened 2026-08-29
fix(setup): Blank Slate must remove installer-seeded bundled skills

juancrfig:
## Summary

Picking **Blank Slate** in first-time setup only wrote the `.no-bundled-skills` opt-out marker and re-ran `sync_skills()`. That prevents *future* seeding but never deleted the skills the installer (`setup-hermes.sh`) already copied in *before* the wizard ran — so the user was left with the entire 80+ skill catalog after explicitly choosing a blank profile.

This explains the recurring report: user selects Blank Slate, then every `hermes update` is blamed for "re-injecting" skills — but updates already honor the marker and re-inject nothing. The skills were never removed in the first place.

## Root cause (reproduced)

1. `setup-hermes.sh` line ~431 runs `sync_skills` and seeds the full catalog into `~/.hermes/skills/` before setup.
2. Selecting Blank Slate wrote the marker + `sync_skills` — which, with the marker present, seeds only the essential `hermes-agent` skill and leaves everything already on disk untouched.

Repro against a temp `HERMES_HOME`: after a full installer-style sync (82 skills + manifest), Blank Slate's old code left **82 skills** on disk. `hermes update` was never the culprit.

## Fix

The Blank Slate path now calls `remove_pristine_bundled_skills()` after writing the marker, then re-syncs. It deletes only pristine, manifest-tracked, unmodified bundled skills (user-modified, hub-installed, and locally-authored skills are preserved), leaving the profile with just the essential `hermes-agent` operating manual.

After the fix, the same repro ends with **1 skill** (`hermes-agent`) — a genuinely blank slate.

## Test Plan

- New regression test `test_finish_now_removes_already_seeded_bundled_skills` verifies the finish-now path writes the marker, calls `remove_pristine_bundled_skills(dry_run=False)`, then re-syncs.
- `pytest tests/hermes_cli/test_setup_blank_slate.py` → 5 passed
- `pytest tests/tools/test_skills_sync.py` → 34 passed
- Manual E2E repro of the installer→Blank-Slate flow on a temp `HERMES_HOME` confirms 82 → 1 skills.

### DIFF
--- a/hermes_cli/setup.py
+++ b/hermes_cli/setup.py
@@ -3602,8 +3602,18 @@ def _run_blank_slate_setup(config: dict, hermes_home, is_existing: bool):
         # `hermes update` runs don't re-inject them. Essential skills (the
         # `hermes-agent` operating manual) are still seeded by the sync.
         try:
-            from tools.skills_sync import set_bundled_skills_opt_out, sync_skills
+            from tools.skills_sync import (
+                remove_pristine_bundled_skills,
+                set_bundled_skills_opt_out,
+                sync_skills,
+            )
             set_bundled_skills_opt_out(True)
+            # The installer seeds the full bundled catalog before this wizard
+            # runs, so opt-out alone is not enough: delete the already-seeded
+            # bundled skills (only pristine, manifest-tracked, unmodified ones
+            # are removed — user/hub/local skills are preserved), then re-sync
+            # to restore just the essential `hermes-agent` operating manual.
+            remove_pristine_bundled_skills(dry_run=False)
             sync_skills(quiet=True)
         except Exception as exc:
             logger.debug("blank-slate skill opt-out error: %s", exc)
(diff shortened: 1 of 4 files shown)


Enough1122:

**Overall:** Fixes Blank Slate leaving 80+ installer-seeded skills on disk.

**What it does**
- `hermes_cli/setup.py:11` + `:31` — both `_run_blank_slate_setup` (finish-now) and `_blank_slate_walkthrough` (opt-out branch) now call `remove_pristine_bundled_skills(dry_run=False)` after `set_bundled_skills_opt_out(True)` and before `sync_skills(quiet=True)`, so only essential skills survive.
- `tools/skills_sync.py:191` — `remove_pristine_bundled_skills` now always skips `_essential_names()` (e.g. `hermes-agent`) with reason `"essential (kept)"`, even when pristine + manifest-tracked.
- Tests `tests/hermes_cli/test_setup_blank_slate.py:66` and `tests/tools/test_skills_sync.py:111` cover opt-out call order and essential preservation.

**Non-blocking notes**
- `setup.py:23` swallows all exceptions with `logger.debug` — acceptable for best-effort wizard, but a user who manually edited a bundled skill will silently keep it (reported via `skipped` in `remove_pristine_bundled_skills` — wizard doesn't surface that count; could log at `info`).
- Essential guard is in the inner manifest loop, so no new filesystem walk added.

No functional issues found.

PR #98029, opened 2026-08-29
fix(config): never persist an expanded secret over a ${VAR} template

dizhaky:
## Summary
`save_config()` wrote a **live unified-gateway bearer token as plaintext** into `~/.hermes/config.yaml` on mfc1, replacing `Bearer ${UNIFIED_GATEWAY_BEARER_TOKEN}` — in a file that was mode **0644**. (DAN-3143)

## Root cause
`_preserve_env_ref_templates` already exists to prevent exactly this. But all three of its checks compare `current` against an expansion of `raw`, and `_expand_env_vars` leaves an unresolved `${VAR}` **verbatim** when the variable isn't in `os.environ`.

On mfc1, `UNIFIED_GATEWAY_BEARER_TOKEN` lives in `~/.hermes/.env` (loaded by the daemon) and never reaches `os.environ` at save time. So the secret arrived already expanded from a source the comparison couldn't reproduce — no branch matched, the guard fell through to `return current`, and the credential was persisted.

Reproduced directly:
```
var UNSET, token came from elsewhere  -> template preserved? False   ← the leak
var SET and matches                   -> template preserved? True
```

## Fix
`_is_unresolved_secret_ref()` fires only when **both** hold:
1. the template names a credential (`TOKEN|SECRET|KEY|PASSWORD|CREDENTIAL|BEARER|AUTH`), **and**
2. that variable is **unset** — so no expansion of the template could have produced the new value; it must have come from elsewhere.

Then keep the template instead of writing the secret.

### The "is unset" half is what makes this safe
My first attempt keyed only on *"does the value look like a secret"* and **broke `test_save_config_allows_intentional_secret_value_change`** — pasting a replacement API key through the UI is a deliberate, caller-owned edit and must persist. It still does: when the var **is** set, the guard doesn't fire. Now pinned by `test_intentional_replacement_wins_when_the_var_is_set`.

## Test plan
5 new cases: the mfc1 leak, rotation between load and save, non-secret literal edit still caller-owned, clearing to empty, and the intentional-replacement boundary.

- Leak case **fails against `main`**, passes here; the other four pass both ways
- **18 passed** across env-ref, secret-prompt and auth-usable suites
- All pre-existing env-ref tests pass unchanged

## Already remediated on mfc1 (out-of-band)
- Literal replaced with the placeholder — verified **byte-identical** to the `~/.hermes/.env` value, so behaviour is unchanged; confirmed it resolves to the full 64-char token at runtime and the gateway still answers **HTTP 200**
- `config.yaml` and its backups now mode **0600**
- Backup kept at `~/.hermes/config.yaml.bak-dan3143-*`

## ⚠️ Still outstanding — not something I should do
**Token rotation.** The credential sat world-readable for an unknown period and appears still valid. Rotating it is a credential operation that needs a human. Tracked in DAN-3143.

Refs DAN-3143


dizhaky:
Closing immediately — opened against the wrong repository by mistake. This is intended for a downstream fork and was never meant for this upstream project. Apologies for the noise.

andrexibiza:
Upstream defect topology is now preserved in #98717. This PR should remain closed/unmerged exactly as its author intended: it was opened against this repository by mistake and its 7,656-file downstream branch is **not** an upstream implementation carrier.

What is being carried forward is the narrow reproduced provenance from this thread: when a raw `${VAR}` secret template is unresolved at save time but the live credential arrived through another runtime authority, `_preserve_env_ref_templates()` can fail to connect that value back to the template and allow plaintext persistence. #98717 records the current-main source boundary, regression shape, intentional-replacement positive control, and acceptance contract while preserving @dizhaky's contribution here.

Issue #97653, opened 2026-08-29
[Feature]: Per-dispatch model/provider override for delegate_task subagents


DavidMetcalfe:
---
name: "✨ [Feature]: Per-dispatch model/provider override for delegate_task subagents"
about: "delegate_task has no per-task model/provider parameter, forcing agents that need different models to rewrite persistent config"
title: "[Feature]: Per-dispatch model/provider override for delegate_task subagents"
labels: ["enhancement"]
---

### Problem or Use Case

`delegate_task` tasks accept `goal`, `context`, `role`, and `toolsets` — but no `model` or `provider` parameter. Subagent routing is fixed per-install by the persistent `delegation.model` / `delegation.provider` config keys.

This leaves an agent that needs a *different* model for one dispatch with exactly one lever: rewriting persistent config with `hermes config set delegation.model ...` — which changes routing for **all future subagents in every session** until someone changes it back, and which (per companion issue: confirm gate for model-routing writes) currently happens with no user-visible notice.

Concrete failure sequence from my install: a session hit `HTTP 400: upstage/solar-pro-4 is not a valid model ID` on a delegation batch, decided the config value was stale, ran `hermes config set delegation.model deepseek/deepseek-v4-flash` — and the persistent config never went back. The user (me) discovered the routing change days later in the settings UI. A per-dispatch override would have made the legitimate response ("re-dispatch on a model that works, leave config alone") possible without any persistent side effect.

It also blocks legitimate *capability* routing that has nothing to do with failures: an agent that knows a task is vision-heavy, or wants a cheap high-throughput model for mechanical extraction and a stronger model for design analysis, currently cannot express that per-task. The only way to get it is to mutate global state.

### Proposed Solution

Add an optional per-task (and/or per-batch) parameter to `delegate_task`:

```json
{
  "tasks": [
    {
      "goal": "...",
      "model": "deepseek-v4-flash",       // optional
      "provider": "deepseek",             // optional
      "toolsets": ["terminal", "file"]
    }
  ]
}
```

Semantics:
- **Unset → current behavior** (inherit `delegation.model`/`delegation.provider`, then parent chain). No behavior change for existing users.
- **Validation**: an explicitly-passed `model`/`provider` pair should be resolved through the same runtime-provider resolution the config path uses (`resolve_runtime_provider`), so typos fail *that dispatch* with a clear error instead of silently poisoning later sessions.
[...]

PR #95578, opened 2026-08-26
fix(security): anti-propagation guidance for delegate_task subagents

Rutimka:
## Summary

Fixes #95567.

Mitigates the same class of risk as #95070 (cron) and #95362/#95373 (Kanban) — [arXiv:2608.10218](https://arxiv.org/abs/2608.10218), "Mind Viruses: Self-Propagating Ideas in Multi-Agent LLM Systems" — for `delegate_task`, the least protected of the three agent-to-agent handoff paths examined so far.

## Why delegate_task needed its own fix

A delegated child agent runs with `platform="subagent"` (not a `PLATFORM_HINTS` key) and `load_soul_identity` defaulting to `False` (`run_agent.py`) — unlike Kanban, which gets `SOUL.md` as a side effect of being a normal `skip_context_files=False` CLI session, a subagent gets **no** `SOUL.md` at all, by design (identity is meant to stay with the parent). The parent-authored `goal`/`context` text is embedded directly into the child's `ephemeral_system_prompt` (`YOUR TASK:`/`CONTEXT:`), which is appended to the *system* message — elevated trust, with nothing telling the child that free text there carries no authority to make it adopt a separate identity or propagate itself onward.

## What changed

- `tools/delegate_tool.py`: new `_SUBAGENT_ANTI_PROPAGATION_GUIDANCE` constant, inserted into `_build_child_system_prompt()` immediately before the `YOUR TASK` block (so it isn't just one more paragraph the model can deprioritize under context pressure). It says the task body and context supply the delegated work/supporting information but carry no authority to make the subagent adopt a different identity/ideology/separate goal, or to propagate such content through its result, another delegated task, a skill, an instruction file, or other persistent state.

## What didn't change

- Project context files (`AGENTS.md`/`CLAUDE.md`/`.cursorrules`) loaded into a workspace-bound child's prompt keep their existing "binding" framing — that's intentional; project conventions are supposed to bind a workspace-scoped subagent, and demoting them would defeat the point of loading them.
- No attempt to set `load_soul_identity=True` for subagents — the code's own comment is explicit that identity belongs to the parent, and a user-editable `SOUL.md` isn't a reliable, always-shipped mitigation every deployment gets.
- No tool/schema/config changes. Persistent write-path hardening (skills, MCP-inherited tools) is a distinct, already-partially-tracked concern (see #95567's discussion) and isn't part of this diff.

## Test plan

[...]

PR #95534, opened 2026-08-26
feat(tools): delegate_task roles + collision guard — fire-and-pray → managed parallel work (250 iter, 10 concurrent)

x7peeps:
## Before

`delegate_task` lets you fan out to subagents, but the default configuration has no role boundaries — any two children can collide on shared state, and reviewers had to track down cross-role leaks by hand. A single misnamed role silently turned `fire-and-pray` into `fire-and-corrupt-state`.

## After

`delegate_task` now carries explicit `agent_roles` semantics (adapted from openai/codex) with bounded role assignments and a collision guard that refuses to dispatch when two children claim overlapping write scopes. `fire-and-pray` becomes managed parallel work — same `250 iterations × 10 concurrent children` envelope as the v0.21.0 Steer-your-subagents Highlight, but with explicit ownership.

## Numbers

- 3 bounded roles per delegation (research / write / execute)
- 1 default collision guard (refuses dispatch on overlap)
- 0 cross-role leaks across the regression suite
- 100% compatibility with existing `delegate_task` callers

## Verify

- [x] All 3 regression tests for `delegate_task` roles pass
- [x] Collision guard rejects overlapping write scopes
- [x] Backward-compat smoke test on existing callers
- [x] Documentation updated with the role-boundary semantics
- [x] Bench: dispatch latency unchanged (within 2% noise vs main)

PR #91252, opened 2026-08-21
feat(delegation): subagents can fork the parent conversation (port of kimi-code#3007)

teknium1:
## Summary

`delegate_task` gains a `fork` parameter: subagents can now start from a one-time snapshot of the parent's conversation instead of a blank context, so work that builds on the current session no longer needs lossy, token-expensive re-briefing. Port of MoonshotAI/kimi-code#3007 (this week's flagship feature in their Agent tool, MIT) adapted to Hermes' delegation architecture.

## Changes

- `tools/delegation_fork.py` (new, ~180 LOC): builds the sanitized parent snapshot —
  - read from the parent's **persisted session transcript** (`get_messages_as_conversation(include_ancestors=True, repair_alternation=True)`), the same durable source the gateway uses for session restore;
  - parent's system prompt stripped (the child builds its own focused prompt); per-row bookkeeping keys (`_row_id`, `display_kind`, …) removed; deep-copied;
  - tail trimmed to the last **completed assistant reply**: the in-flight `delegate_task` tool_call (which can never close inside the child), dangling tool results, and the triggering user prompt are removed so role alternation and tool-call adjacency stay wire-valid — mirrors kimi-code's mid-turn trailing-message handling;
  - capped to `delegation.fork_max_messages` most-recent messages (default 200, floor 10), with post-cut boundary repair;
  - kickoff goal framed with an inheritance notice (snapshot = parent's reference material, not the child's own actions) — kimi-code's framing, adapted.
- `tools/delegate_tool.py`: `fork` kwarg end-to-end (registry handler → `delegate_task` → child attachment → `_run_single_child` seeds via `run_conversation(conversation_history=...)` — the existing gateway-restore parameter, no new injection path). Top-level `fork` is the batch default; per-task `fork` overrides. Missing session DB / empty transcript degrades to a blank-context spawn with a log line. Schema documents the cost trade-off so models keep fork off for independent tasks.
- Docs in the same PR: `features/delegation.md` (new "Forked Subagents" section, warning-box update) and `configuration.md` (`delegation.fork_max_messages`).
- `tests/tools/test_delegation_fork.py`: 15 tests — tail trimming (dangling call, completed round-trip, trailing user prompt), system/bookkeeping stripping, cap + post-cut repair, deep-copy isolation, degrade paths (no DB, empty history, DB error), schema/signature wiring.

## Validation

| | Before | After |
|---|---|---|
[...]

PR #98031, opened 2026-08-29
feat(delegation): per-task model selection and profile identity for delegate_task

CaptainPickard:
## Summary

- **Per-task model selection**: each task in a `delegate_task` batch can name a model ("opus", "gpt-5", "glm", or a full "vendor/model" slug), resolved leniently through the existing `model_switch` pipeline (same as `/model`). Provider is resolved, not dictated.
- **Per-task profile identity**: each task can name a Hermes profile; the child loads that profile's `SOUL.md`, `IDENTITY.md`, and `AGENTS.md` as its system prompt, and reads model/provider from its `config.yaml` when not explicitly overridden. The child becomes the named bot rather than a generic subagent.
- Both features are **config-gated** (default off) so the existing subagent contract is unchanged when the flags are off. Schema fields only appear when the corresponding flag is enabled.

## Motivation

Multi-role workflows (e.g. dispatching a code reviewer, test validator, and security auditor in one `delegate_task` batch) currently require every child to run on the same model with the same generic identity. This PR adds two independent opt-in capabilities:

1. **Model routing** lets the agent fan work across different models in a single batch -- useful for comparing results across models, or routing specialized work to specialized models.

2. **Profile identity** lets each child adopt a named profile's identity files and model config, so a delegation batch can dispatch specialized bots with their own personas, rules, and model configurations.

Both flags are off by default because:
- Per-task model routing can send work to a more expensive model than expected.
- Loading arbitrary profile files into a child prompt is a trust boundary that should be deliberate.
- The schema fields only appear when opted in, keeping the tool surface minimal otherwise.

The model selection feature overlaps conceptually with the `kilocode-port/per-task-delegation-model` branch, but this PR takes a different approach: it adds profile identity as a complementary feature and uses a unified code path. Happy to coordinate if there's interest in merging the approaches.

## Changes

- `tools/delegate_tool.py`:
  - `_get_allow_model_selection()` / `_get_allow_profile_identity()`: config-gated flag getters
  - `_resolve_task_model_creds()`: resolves a per-task model name via `model_switch.switch_model()` (reuses the `/model` command's resolution pipeline)
  - `_load_profile_identity()`: loads a named profile's `SOUL.md`, `IDENTITY.md`, `AGENTS.md`, and `config.yaml` model/provider
  - `_build_child_system_prompt()`: accepts optional `profile_identity` to replace the generic subagent preamble with the profile's identity files
  - `_build_child_agent()`: passes `profile_identity` through to the system prompt builder
  - `delegate_task()`: accepts `model` and `profile` params; per-task resolution in the child loop (gated)
  - `_build_dynamic_schema_overrides()`: adds `model`/`profile` fields only when the corresponding flag is on; deep-copies tasks schema to avoid mutating the static schema
  - Registry handler passes `model`/`profile` through
- `hermes_cli/config_defaults.py`: adds `allow_model_selection` and `allow_profile_identity` defaults (both `False`)
- `website/docs/user-guide/features/delegation.md`: documents both features with config examples
- `tests/tools/test_delegate_per_task_overrides.py`: 17 tests covering schema gating, flag getters, model resolution, profile identity loading, and system prompt construction

## Test Plan

- [x] 17 new tests pass (`python3 -m pytest tests/tools/test_delegate_per_task_overrides.py -v`)
- [x] 70 existing delegate tests pass (no regressions: `python3 -m pytest tests/tools/test_delegate.py -v`)
- [x] Syntax checks pass for all modified files
- [ ] CI passes on GitHub Actions

## Notes for Reviewers

- The flag-off path is byte-identical to prior behavior -- no schema changes, no code path changes when both flags are `False`.
- The `_resolve_task_model_creds` function reuses the existing `model_switch.switch_model()` pipeline rather than reinventing model resolution, so aggregator-aware aliasing, fuzzy matching, and cross-provider fallback all come for free.
- The profile identity feature is the unique contribution not present in the `kilocode-port/per-task-delegation-model` branch. Both features can be combined: a task can name both `model` and `profile`, with the explicit model taking precedence over the profile's config.
- The dynamic schema override deep-copies the tasks property before adding per-task fields, so `DELEGATE_TASK_SCHEMA` is never mutated.

### DIFF
--- a/hermes_cli/config_defaults.py
+++ b/hermes_cli/config_defaults.py
@@ -2118,6 +2118,25 @@ DEFAULT_CONFIG = {
                                      # (floor 30s) to enforce a hard cap.
         "reasoning_effort": "",  # subagent effort: "ultra", "max", "xhigh", "high",
                                  # "medium", "low", "minimal", "none" (empty = inherit)
+        # Allow the agent to pick a per-task model when fanning out work, e.g.
+        # comparing the same task across several models. When True, each
+        # delegate_task entry may carry a `model` field naming a model
+        # ("opus", "gpt-5", "glm") — resolved leniently against the parent's
+        # current provider (aggregator-aware), falling back to delegation.provider.
+        # Off by default: the documented contract is that subagents inherit the
+        # parent model, and per-task model selection can route work to a more
+        # expensive model than the user expects. Opt in deliberately.
+        "allow_model_selection": False,
+        # Allow the agent to name a per-task Hermes profile when fanning out
+        # work. When True, each delegate_task entry may carry a `profile` field
+        # naming a profile under the Hermes profiles root. The child loads that
+        # profile's SOUL.md, IDENTITY.md, and AGENTS.md as its system prompt
+        # identity, and reads model/provider from its config.yaml when not
+        # explicitly overridden — so the child "becomes" the named bot rather
+        # than a generic subagent. Off by default: loading arbitrary profile
+        # files into a child prompt is a trust boundary that should be
+        # deliberate.
+        "allow_profile_identity": False,
         "max_concurrent_children": 10,  # unified concurrency cap: max parallel children per batch
                                        # AND max concurrent background (background=true)
                                        # delegation units. New async dispatches beyond the cap
--- a/tests/tools/test_delegate_model_selection.py
+++ b/tests/tools/test_delegate_model_selection.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+"""
+Integration tests for per-task model selection in delegate_task.
+
+Mirrors the test file name and structure from Kilo-Org/kilocode#11786
+(upstream kilocode-port/per-task-delegation-model branch). These tests
+exercise _resolve_task_model_creds with REAL switch_model() calls rather
+than mocks, complementing the mocked unit tests in
+test_delegate_per_task_overrides.py.
+
+Run with:  python -m pytest tests/tools/test_delegate_model_selection.py -v
+"""
+
+import unittest
+from unittest.mock import patch
+
+from tools.delegate_tool import (
+    DELEGATE_TASK_SCHEMA,
+    _build_dynamic_schema_overrides,
+    _get_allow_model_selection,
+    _resolve_task_model_creds,
+)
+
+
+class _FakeParent:
+    """Minimal parent agent for credential anchoring."""
+
+    provider = "openrouter"
+    model = "anthropic/claude-opus-4.8"
+    base_url = "https://openrouter.ai/api/v1"
+    api_key = "sk-test"
+
+
+_BASE_CREDS = {
+    "model": None,
+    "provider": None,
+    "base_url": None,
+    "api_key": None,
+    "api_mode": None,
+    "command": None,
+    "args": None,
+}
+
+
+class TestSchemaGating(unittest.TestCase):
+    """The per-task model field only appears when the flag is enabled."""
+
+    def test_flag_off_no_model_field(self):
+        with patch("tools.delegate_tool._load_config", return_value={}):
+            ov = _build_dynamic_schema_overrides()
+        props = ov["parameters"]["properties"]
+        self.assertNotIn("model", props)
+        self.assertNotIn("model", props["tasks"]["items"]["properties"])
+
+    def test_flag_on_adds_model_field(self):
+        with patch(
+            "tools.delegate_tool._load_config",
+            return_value={"allow_model_selection": True},
+        ):
+            ov = _build_dynamic_schema_overrides()
+        props = ov["parameters"]["properties"]
+        self.assertIn("model", props)
+        self.assertEqual(props["model"]["type"], "string")
+        self.assertIn("model", props["tasks"]["items"]["properties"])
+
+    def test_static_schema_never_mutated(self):
+        """Dynamic overrides must not leak the model field into the static schema."""
+        with patch(
+            "tools.delegate_tool._load_config",
+            return_value={"allow_model_selection": True},
+        ):
+            _build_dynamic_schema_overrides()
+        static_props = DELEGATE_TASK_SCHEMA["parameters"]["properties"]
+        self.assertNotIn("model", static_props)
+        self.assertNotIn(
+            "model", static_props["tasks"]["items"]["properties"]
+        )
+
+
+class TestFlagGetter(unittest.TestCase):
+    def test_default_off(self):
+        with patch("tools.delegate_tool._load_config", return_value={}):
+            self.assertFalse(_get_allow_model_selection())
+
+    def test_truthy_on(self):
+        with patch(
+            "tools.delegate_tool._load_config",
+            return_value={"allow_model_selection": True},
+        ):
+            self.assertTrue(_get_allow_model_selection())
+
+
+class TestModelResolutionIntegration(unittest.TestCase):
+    """Integration tests with real switch_model() calls.
+
+    These mirror the kilocode branch's test approach: exercise the actual
+    model_switch pipeline rather than mocking it. Tests may be skipped if
+    the model_switch module cannot resolve models in the test environment
+    (e.g. no provider config available).
+    """
+
+    def test_empty_name_returns_base_unchanged(self):
+        out = _resolve_task_model_creds("", _FakeParent(), _BASE_CREDS)
+        self.assertIs(out, _BASE_CREDS)
+
+    def test_base_creds_not_mutated(self):
+        before = dict(_BASE_CREDS)
+        try:
+            _resolve_task_model_creds("sonnet", _FakeParent(), _BASE_CREDS)
+        except (ValueError, Exception):
+            # Resolution may fail in test env; mutation check is independent
+            pass
+        self.assertEqual(_BASE_CREDS, before)
+
+    def test_full_slug_passthrough(self):
+        """A full vendor/model slug should pass through resolution."""
+        try:
+            out = _resolve_task_model_creds(
+                "openai/gpt-5.4", _FakeParent(), _BASE_CREDS
+            )
+            self.assertEqual(out["model"], "openai/gpt-5.4")
+        except (ValueError, Exception):
+            self.skipTest("model_switch cannot resolve in this environment")
+
+    def test_unresolvable_name_raises(self):
+        with self.assertRaises(ValueError):
+            _resolve_task_model_creds(
+                "zzznotarealmodel-xyz-123", _FakeParent(), _BASE_CREDS
+            )
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/tools/test_delegate_per_task_overrides.py
+++ b/tests/tools/test_delegate_per_task_overrides.py
@@ -1,0 +1,300 @@
+#!/usr/bin/env python3
+"""
+Tests for per-task model selection and profile identity in delegate_task.
+
+Two opt-in features, both gated behind config flags (default off):
+
+1. **Per-task model selection** (``delegation.allow_model_selection``):
+   The agent names a model per task ("opus", "gpt-5", "glm") when fanning
+   work out; resolution reuses the shared ``model_switch`` pipeline so names
+   are matched leniently and the provider is resolved (not dictated).
+
+2. **Per-task profile identity** (``delegation.allow_profile_identity``):
+   The agent names a Hermes profile per task; the child loads that profile's
+   SOUL.md, IDENTITY.md, and AGENTS.md as its system prompt identity, and
+   reads model/provider from its config.yaml when not explicitly overridden.
+   The child "becomes" the named bot rather than a generic subagent.
+
+Both flags are off by default to preserve the "subagents inherit the parent
+model and use a generic identity" contract. The schema fields only appear
+when the corresponding flag is enabled (keeping the tool surface minimal).
+
+Run with:  python3 -m pytest tests/tools/test_delegate_per_task_overrides.py -v
+"""
+
+import os
+import tempfile
+import unittest
+from unittest.mock import patch, MagicMock
+
+from tools.delegate_tool import (
+    DELEGATE_TASK_SCHEMA,
+    _build_dynamic_schema_overrides,
+    _get_allow_model_selection,
+    _get_allow_profile_identity,
+    _resolve_task_model_creds,
+    _load_profile_identity,
+    _build_child_system_prompt,
+)
+
+
+class _FakeParent:
+    """Minimal parent agent for credential anchoring."""
+
+    provider = "openrouter"
+    model = "anthropic/claude-opus-4.8"
+    base_url = "https://openrouter.ai/api/v1"
+    api_key = "sk-test"
+
+
+_BASE_CREDS = {
+    "model": None,
+    "provider": None,
+    "base_url": None,
+    "api_key": None,
+    "api_mode": None,
+    "command": None,
+    "args": None,
+}
+
+
+# ---------------------------------------------------------------------------
+# Schema gating: fields only appear when the flag is on
+# ---------------------------------------------------------------------------
+
+class TestSchemaGating(unittest.TestCase):
+    """Per-task fields only appear when the corresponding flag is enabled."""
+
+    def test_both_flags_off_no_extra_fields(self):
+        with patch("tools.delegate_tool._load_config", return_value={}):
+            ov = _build_dynamic_schema_overrides()
+        props = ov["parameters"]["properties"]
+        self.assertNotIn("model", props)
+        self.assertNotIn("profile", props)
+        task_props = props["tasks"]["items"]["properties"]
+        self.assertNotIn("model", task_props)
+        self.assertNotIn("profile", task_props)
+
+    def test_model_flag_on_adds_model_field(self):
+        with patch(
+            "tools.delegate_tool._load_config",
+            return_value={"allow_model_selection": True},
+        ):
+            ov = _build_dynamic_schema_overrides()
+        props = ov["parameters"]["properties"]
+        self.assertIn("model", props)
+        self.assertEqual(props["model"]["type"], "string")
+        self.assertIn("model", props["tasks"]["items"]["properties"])
+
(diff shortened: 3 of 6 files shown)


Enough1122:

**Overall:** Well-gated feature — per-task model selection and profile identity for `delegate_task`, both off by default.

**What it does**
- `hermes_cli/config_defaults.py:17` + `:28` add `delegation.allow_model_selection` / `allow_profile_identity` (default `False`) with clear trust/cost rationale in comments/docs.
- `tools/delegate_tool.py:346` + `:358` — flag getters; `_resolve_task_model_creds` (`tools/delegate_tool.py:370`) reuses `model_switch.switch_model` anchored on parent provider/creds; raises `ValueError` on unresolvable name instead of silent fallback.
- `tools/delegate_tool.py:444` — `_load_profile_identity` loads `SOUL.md`/`IDENTITY.md`/`AGENTS.md` + `config.yaml` model/provider best-effort; `_build_child_system_prompt` (`tools/delegate_tool.py:524`) replaces generic preamble when `soul` or `identity` present.
- `tools/delegate_tool.py:605` + `:628` — per-task `model`/`profile` resolved inside `delegate_task` loop; `task_creds` threaded through `_build_child_preserving_parent_tools`; dynamic schema (`tools/delegate_tool.py:706`) only advertises fields when flags on (prompt-cache stable).
- Docs `website/docs/user-guide/features/delegation.md:780` explain both flags.

**Non-blocking / consider before merge**
- **Path traversal:** `_load_profile_identity` does `get_default_hermes_root() / "profiles" / profile_name` with raw `profile_name.strip()` (`tools/delegate_tool.py:453`). A task value like `../../.ssh` or `foo/../../other` could escape the profiles root via `Path` joining. Consider validating `profile_name` against `^[A-Za-z0-9_-]+$` and rejecting slashes/dots, or resolving and checking `is_relative_to(profiles_root)`.
- Profile `config.yaml` model fallback only applies when `allow_model_selection` is also true (`tools/delegate_tool.py:642`) — intentional but worth calling out in docs (already noted).
- `_build_dynamic_schema_overrides` deep-copies `_tasks_prop` before mutation (`tools/delegate_tool.py:711`) — good; static schema immutability tested (`tests/tools/test_delegate_per_task_overrides.py:152`).

*Please use your judgment — the traversal guard is the only blocking-adjacent item.*

andrexibiza (commented):
I reviewed exact head `4c37a70d976c8b2a9144fd6e412f1fc791c1214c` against current `main@4f22543509d1b91dc45bcb369447126c5eb14fb7`, including the full 16-file diff / 9-commit history, the behavior-critical child-construction and session-DB paths, the new tests, exact-head check state, and the overlapping delegation/profile/session PRs.

There is good engineering in this branch. In particular, the earlier traversal finding was actually absorbed; model switching now anchors on the effective delegation provider rather than blindly on the parent; cross-provider switches clear stale ACP/pin transport fields; and the model/profile schema remains default-off. Those are real improvements.

I do not think this exact object is safe to land yet, though. The remaining blockers are architectural/correctness issues rather than polish.

### P1 — batch override/profile preflight is not atomic; a later bad task can orphan earlier constructed children

At this head, `delegate_task()` resolves each task's model/profile and then immediately constructs that child in the same `for i, t in enumerate(task_list)` loop (`tools/delegate_tool.py`, around the new per-task routing block at ~4246 and `_build_child_preserving_parent_tools` at ~4332). A later task can then fail model resolution, profile lookup, allowlist validation, or profile-model resolution and return `tool_error(...)` after earlier children already exist.

That is load-bearing because child construction is not inert: `_build_child_agent()` opens a dedicated `SessionDB` and registers the child on `parent_agent._active_children`. The early return does not run those already-constructed children through the normal lifecycle/cleanup path.

There is already a narrower concurrent implementation of this exact slice in #98128 / #97653. Its two-pass preflight is explicit about the same failure mode: resolve every task before any child construction because construction opens the DB handle and registers `_active_children`. That PR is overlap, not a duplicate of this whole profile-identity feature, but this branch needs the same atomic boundary regardless of which vehicle survives.

**Required fix:** preflight *all* model/profile inputs for the whole batch before constructing task 0: profile syntax/existence/allowlist, effective profile provider+model, explicit task model/provider, and the final credential bundle. Then build children from immutable pre-resolved task plans. Add a public-path regression where task 0 is valid and task 1 is invalid and assert zero child constructors ran / `_active_children` is unchanged (and no child DB ownership is left behind).

### P1 — the selected profile's configured provider is loaded but never applied

`_load_profile_identity()` reads both:

- `model.default` → `result["model"]`
- `model.provider` → `result["provider"]`

and the PR/schema promise that a profile's model/provider config is used when not explicitly overridden.

But the dispatch path only consumes `_profile_identity["model"]`:

```py
_profile_model = _profile_identity.get("model")
if _profile_model and allow_model_selection:
    task_creds = _resolve_task_model_creds(_profile_model, parent_agent, creds)
```

`_profile_identity["provider"]` is never used. `_resolve_task_model_creds()` then anchors on `base_creds["provider"]`, which is the delegation/parent route, not the selected profile route.

So a parent/delegation running on provider A can select a profile configured for provider B and still resolve that profile's model against A. That can mean the wrong endpoint/credential/billing route or a false model-resolution failure. The loader tests prove the provider string was read, but not that the child actually receives it.

**Required fix:** resolve a selected profile's provider+model as one routing unit through the canonical runtime resolver, with the documented precedence (`explicit task override > profile config > delegation config > parent`). Add a dispatch-level regression where parent/delegation provider != profile provider and assert the child gets the expected provider, model, base URL, API key/API mode, and transport fields—not merely the loader dictionary.

### P1 — the new model-facing `profile` input still has a known filesystem exception path

Commit `51ea414ba4abbdc8b19f796168b074fc6d8aef1f` records the remaining bug directly: a regex-valid profile name longer than the filesystem component limit can raise `OSError: ENAMETOOLONG` at `profile_path.is_dir()` because that probe sits outside the read-side exception handling.

I would not defer that as a follow-up. This PR is the code introducing a model-facing profile-name → filesystem path boundary, so rejecting hostile/invalid names belongs to this boundary. #90331 is adjacent profile-management work and already uses Hermes' canonical `normalize_profile_name` / `validate_profile_name` path rather than defining a second name law.

**Required fix:** reuse the canonical profile-name validation/bounds (or equivalently make the filesystem probe fail closed) before touching disk, and add a `delegate_task` regression for an overlong alphanumeric name. It should return a bounded tool error and construct nothing.

### P1 — provenance/ownership needs recomposition; two substantial slices already have open upstream owners

The 9-commit branch has grown well beyond the model/profile feature in the PR description. Two of those additions are already open upstream as their own reviewable objects:

- #91252 (`teknium1`, exact head `7e60ebc5d495efa1fba8c25201727c54792ca3cb`) is the existing `fork` implementation porting MoonshotAI/kimi-code#3007. This branch's fork commit explicitly ports that upstream head and carries the same module/tests/schema behavior. That slice is duplicate/derived work; #91252 remains the original repository owner.
- #91271 (`teknium1`, exact head `037ae2724c99d354392a14d1aff92e63e4755f3a`) is the existing per-delegation usage ledger inspired by Copilot CLI. This branch carries that ledger plus a useful `profile` attribution extension. The profile column is complementary work; the underlying ledger is not new ownership.
- #98128 / #97653 (`DavidMetcalfe`) directly overlaps the per-dispatch model/provider routing slice and already supplies the atomic preflight this branch is missing. It is not a duplicate of profile identity, but the two cannot land independently in `delegate_tool.py` without choosing/recomposing one routing owner.

Please preserve those edges and authorship rather than landing rewritten copies as a single omnibus commit train. The clean shape is to drop/rebase onto the surviving upstream owners (or cherry-pick the original objects with authorship intact) and keep only this PR's genuinely complementary profile-identity delta. External source credit to Kilo, Moonshot/Kimi, and Copilot CLI should remain as well.

There are additional *adjacent/complementary* edges worth preserving during that rebase rather than flattening them into duplicates: #95534 (bounded reusable roles/personas), #95578 (delegate anti-propagation prompt guidance), and #92001 (explicit profile-home ownership for delegation persistence). In particular, if #95578 lands first, the profile-identity prompt path must not replace/drop its security guidance; if #92001 lands first, new child persistence must reuse its explicit-home boundary rather than recreate ambient storage selection.

### P1 landing gate — this exact object has no hosted acceptance and is far off current main

Current topology at review time:

- PR head: `4c37a70d976c8b2a9144fd6e412f1fc791c1214c`
- current main: `4f22543509d1b91dc45bcb369447126c5eb14fb7`
- merge base: `1e21fe862421081e14af4564076af74bc58c050b`
- branch: **9 ahead / 154 behind**, `mergeable=false`
- exact head: **0 workflow runs, 0 check runs, 0 commit statuses**

The local targeted tests are useful evidence, but they cannot substitute for hosted acceptance of the submitted object—especially after the branch has accreted fork, usage, and session-persistence behavior not reflected in the original test-plan summary. Recompose on current main, then reacquire exact-head CI/Docker/Nix (and every submitted commit must be green, not only a merge ref/head snapshot).

The size/ownership gate also still fails: this PR edits `tools/delegate_tool.py` past ~5.5K lines, `cli.py` around ~14K, and `agent/agent_init.py` past ~2.9K. New routing/identity/persistence authority should move behind bounded owners rather than add another concern to those godfiles. `tools/delegation_fork.py` is the right *shape* of extraction; it just already belongs to #91252's lineage.

### Separate the session-title side channel from this feature

The late session-title commits are also orthogonal to per-task model/profile identity and currently write `sessions.title` with raw `sqlite3` from `delegate_tool.py`, including a fallback that can synthesize a profile DB path and finally falls back to `/home/hermeswebui/.hermes/profiles/io/state.db`.

That creates a second persistence authority beside `SessionDB` exactly while session/profile ownership is being hardened elsewhere (#92001 and the broader session-lifecycle work). The pre-run UPDATE is also necessarily best-effort because the row may not exist yet, then `_finalize_child_results` repeats the write.

I would remove that ride-along from this PR and give descriptive subagent titles their own canonical session-owner change, or route it through the existing session title API with real lifecycle/profile-home tests. It should not be coupled to whether profile identity lands.

### What I would re-review

A much cleaner next object would be: current-main rebase; fork and usage work removed/recomposed onto #91252/#91271 with original credit; one owner selected for the #97653 model-routing slice; all task routes/profile identities preflighted before any child construction; profile provider actually honored; canonical profile-name validation; bounded module ownership; and fresh exact-object hosted CI for every commit.

The core product idea is useful, and the conservative defaults plus the resolver hardening are solid. The branch now needs the same care at the *composition* boundary that it already applies inside the resolver. Once those ownership and atomicity seams are closed, this will be much easier to reason about and much safer to merge.

CaptainPickard:
Thanks for the thorough review. The bug findings are well-taken and we agree with the composition direction. Here is our plan.

## What we are doing: full rebase + recompose

We will rebase on current `main` (4f22543) and slim the branch down to only the profile-identity delta. The fork, usage-ledger, and session-title commits will be removed from this PR.

### Commits being dropped from this PR

- `35aca686f2` feat(delegate): fork context parameter — belongs to #91252 (teknium1). We will not carry a duplicate.
- `f71f2e1d12` feat(delegate): per-delegation usage ledger with profile column — the ledger belongs to #91271 (teknium1). The profile attribution column is complementary; we will offer it as a follow-up patch on top of #91271 rather than bundling the whole ledger here.
- `f11601711b` + `4c37a70d97` session-title side channel (raw sqlite3 writes, synthesized profile DB path fallback) — removed entirely. Descriptive subagent titles should go through the canonical session owner, not a second persistence authority in delegate_tool.py. We will separate that into its own change with proper lifecycle/profile-home tests.

### Commits staying (the profile-identity delta)

- `f43cdce6c0` per-task model selection and profile identity
- `52c0861e9f` profile_name path traversal guard (will be replaced with canonical `validate_profile_name` — see P1 #3 below)
- `3d535fd8c0` docs convergence
- `51ea414ba4` test gap closure (will be updated for the new code paths)

## Fixes for each P1

### P1 #1 — atomic batch preflight

Agreed. The current loop constructs children inline with resolution, so a later failure orphans earlier children that already have open SessionDB handles and `_active_children` registrations.

**Fix:** Two-pass design. Pass 1 resolves every task's effective model, provider, profile identity, allowlist status, and credential bundle into an immutable plan list. If any task fails resolution, return a bounded tool error with zero child construction. Pass 2 iterates the pre-resolved plans and constructs children.

We will add a regression test where task 0 is valid and task 1 has an invalid profile, then assert `_active_children` is unchanged and no child DB handle was opened.

We will also review #98128 (DavidMetcalfe) and coordinate so we do not land conflicting routing owners in `delegate_tool.py`. If #98128 lands first, we will rebase onto its preflight boundary rather than reinventing one.

### P1 #2 — profile provider loaded but never applied

Confirmed bug. `_load_profile_identity()` reads `model.provider` but the dispatch path only consumes `model`, then `_resolve_task_model_creds()` anchors on `base_creds["provider"]` (the parent/delegation route) instead of the profile's route.

**Fix:** Resolve a selected profile's provider+model as one routing unit through the canonical resolver. Precedence: explicit task override > profile config > delegation config > parent. The preflight pass will produce the final `(provider, model, base_url, api_key, transport fields)` tuple per task, not just the model string.

**Regression test:** parent/delegation provider A, profile configured for provider B, no explicit task override. Assert the child receives provider B's endpoint, credentials, and transport fields, not A's.

### P1 #3 — ENAMETOOLONG filesystem exception

Agreed that name-to-filesystem-path validation belongs to this boundary since this PR introduces the model-facing profile name input.

**Fix:** Replace the ad-hoc regex guard with Hermes' canonical `normalize_profile_name` / `validate_profile_name` (the path #90331 uses). The validation runs in the preflight pass before any disk probe. Overlong, hostile, or invalid names return a bounded tool error and construct nothing.

**Regression test:** delegate with an overlong alphanumeric profile name, assert bounded error, zero child construction.

### P1 #4 — provenance/ownership

Acknowledged. The fork slice is #91252's work (teknium1, porting kimi-code#3007). The usage ledger is #91271's work (teknium1, Copilot CLI port). We are dropping both from this PR (see above).

For the #97653/#98128 model-routing overlap (DavidMetcalfe): #97653 is the feature issue, #98128 is the PR. We will coordinate on a single routing owner rather than landing parallel implementations. If #98128 lands first, we rebase onto it. If ours lands first, we will ensure the preflight boundary is compatible.

Adjacent edges we will respect during rebase:
- #95534 (bounded roles) — our profile identity will not redefine roles/personas.
- #95578 (anti-propagation prompt guidance) — our prompt path will not replace or drop its security guidance. If #95578 lands first, we will build on top of it.
- #92001 (profile-home ownership for delegation persistence) — new child persistence will reuse its explicit-home boundary rather than recreating ambient storage selection.

### P1 #5 — landing gate

Agreed. After the rebase, we will push the exact head and verify hosted CI/Docker/Nix checks run green on every commit, not just the head snapshot. The branch will be current-main-based with a clean mergeable state before requesting re-review.

On the godfile concern: we agree `delegate_tool.py` is too large. `tools/delegation_fork.py` was the right extraction shape, but it belongs to #91252's lineage. For this PR's scope, we will keep the profile-identity logic behind bounded helpers and avoid growing delegate_tool.py further where possible. A broader extraction of routing authority into its own module is worth doing, but we will scope that as a follow-up rather than expanding this PR.

## Timeline

1. Rebase on current main (drop fork, usage, session-title commits).
2. Implement atomic two-pass preflight (P1 #1).
3. Fix profile provider routing (P1 #2).
4. Replace ad-hoc name validation with canonical path (P1 #3).
5. Update and add regression tests for all three fixes.
6. Force-push the clean branch, verify hosted CI green on every commit.
7. Request re-review.

Appreciate the review. The core resolver hardening and default-off schema will carry forward into the clean object.

CaptainPickard:
Updated branch pushed. Summary of changes since the review:

## Branch recomposed

- Rebased on current `main` (4f22543). The branch is now 5 commits, 6 files, +1836/-18, `mergeable=true`.
- Fork context parameter, usage ledger, and session-title side channel removed. Those slices belong to #91252, #91271, and a separate session-owner change respectively.

## P1 fixes applied (commit 7d1ad50f8c)

**P1 #1 -- atomic preflight:** The dispatch loop is now two-pass. Pass 1 resolves every task's role, model, profile identity, profile model fallback, and credential bundle into immutable per-task plan dicts. Any resolution failure returns `tool_error` with zero children constructed. Pass 2 iterates the pre-resolved plans and constructs children -- no model resolution or profile loading happens in this pass. Regression test: task 0 valid + task 1 invalid profile -> error names task 1, `_build_child_preserving_parent_tools` called 0 times, `_active_children` unchanged.

**P1 #2 -- profile provider routing:** New `_resolve_profile_model_creds()` resolves a profile's model+provider as one routing unit through `resolve_runtime_provider`, anchored on the profile's configured provider (not the parent's). The profile's `base_url`, `api_key`, `api_mode`, and `request_overrides` all come from the profile's provider. Stale ACP/pin fields are cleared on cross-provider switches. Precedence: explicit task model > profile config (model+provider) > delegation config > parent. Regression test: parent on provider A, profile configured for provider B -> child receives provider B's endpoint and credentials.

**P1 #3 -- ENAMETOOLONG guard:** New `_validate_profile_name()` checks `^[A-Za-z0-9_-]+$` regex and length <= 255 before any filesystem access. `profile_path.is_dir()` is wrapped in try/except OSError as defense-in-depth. Regression test: 256-char alphanumeric name returns None, no OSError.

**P1 #4 -- profile name propagation:** `_profile_name` added to `_load_profile_identity()` return dict. `child._delegate_profile` is set in `_build_child_agent` from the identity dict's validated name.

## Ownership

- Fork slice dropped (belongs to #91252).
- Usage ledger dropped (belongs to #91271).
- Session-title side channel dropped (will be a separate session-owner change).
- Model-routing slice: will coordinate with #98128 if it lands first.
- Adjacent edges respected: #95534, #95578, #92001.

## Test results

54 passed, 0 failures across `test_delegate_per_task_overrides.py`, `test_delegate_model_selection.py`, `test_delegate_test_gap.py`. 4 new regression tests added. Import check passes. No fork/usage/session-title references in the P1 commit.

CI is pending on the new head. Will confirm once checks complete.

PR #98041, opened 2026-08-29
fix(desktop): keep long task lists scrollable

### LINKED ISSUE #97997: [Bug][Desktop] Tasks list stops rendering after 18 items and cannot scroll down
## Summary

Hermes Desktop's active goal/task list becomes unusable once more than 18 tasks are present. With `Tasks 18/20`, the visible list ends around item 18; tasks 19 and 20 are not accessible, and mouse-wheel/trackpad scrolling does not move the list downward.

## Environment

- Hermes Agent: v0.20.6 (`v2026.8.3-5500-ge60983a697`)
- Platform: macOS on Apple Silicon
- Surface: Hermes Desktop active goal/task panel
- Example: 20 tasks in the active goal (`Tasks 18/20`)

## Steps to reproduce

1. Open Hermes Desktop.
2. Start or resume a goal with at least 20 tasks.
3. Open the active goal/task list.
4. Try to scroll down after the 18th visible task.

## Expected behavior

The list should be vertically scrollable so every task is reachable. Mouse-wheel/trackpad and keyboard scrolling should work, and the viewport should expose an obvious scroll affordance when more tasks exist below it.

## Actual behavior

Only the first approximately 18 tasks can be viewed. Scrolling down has no effect, so later tasks cannot be inspected or selected. The bottom composer overlaps/occludes the remaining list area.

## Evidence

The reporter's screenshot shows `Tasks 18/20`; the task list is cut off at the bottom by the composer, with no usable downward scroll.

## Impact

Tasks beyond the viewport are effectively hidden, making long goals impossible to monitor or manage from Desktop.

This may be related to the task-panel/container height or overflow handling when the composer is docked at the bottom.

sylvesterkaczmarek:
## Summary

- keep the composer status card from flex-shrinking inside the capped status stack
- let the outer `overflow-y-auto` container own overflow so every task row remains reachable
- add a focused regression test for the non-shrinking dock-card contract

## Root cause

The status stack is a capped flex column with `overflow-y-auto`, but its inner dock card was still a shrinkable flex child and also used `overflow-hidden`. With enough task rows, flexbox shrank that card to the height cap and clipped the later rows inside it, so the outer container had no overflow left to scroll.

Adding `shrink-0` keeps the card at its content height, which makes the existing outer scroll container work as intended.

## Testing

- added `composer-dock.test.ts` covering the non-shrinking card contract
- tests not run locally

Closes #97997

### DIFF
--- a/apps/desktop/src/components/chat/composer-dock.test.ts
+++ b/apps/desktop/src/components/chat/composer-dock.test.ts
@@ -1,0 +1,9 @@
+import { describe, expect, it } from 'vitest'
+
+import { composerDockCard } from './composer-dock'
+
+describe('composerDockCard', () => {
+  it('does not shrink inside the capped status-stack scroll container', () => {
+    expect(composerDockCard('top').split(/\s+/)).toContain('shrink-0')
+  })
+})
--- a/apps/desktop/src/components/chat/composer-dock.ts
+++ b/apps/desktop/src/components/chat/composer-dock.ts
@@ -19,9 +19,13 @@ const composerDockEdge = (edge: 'bottom' | 'top') =>
 
 /** Glassy docked card — the status stack / queue. Paints the SAME
  *  `--composer-fill` as the surface, so rest / scrolled / focused / drawer-open
- *  all match the composer by construction. */
+ *  all match the composer by construction.
+ *
+ * Keep the card non-shrinking inside capped flex scroll containers. Otherwise
+ * flexbox compresses the card to the cap and its own overflow-hidden clips
+ * later status rows before the outer status stack gets anything to scroll. */
 export const composerDockCard = (edge: 'bottom' | 'top' = 'top') =>
-  cn(composerDockEdge(edge), composerFill, composerSurfaceGlass)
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Minimal, correct flexbox fix for long task lists.

**What it does**
- `apps/desktop/src/components/chat/composer-dock.ts:32` adds `shrink-0` to `composerDockCard` so the capped `overflow-y-auto` status stack, not the inner dock card, owns scrolling. Previously the card's `overflow-hidden` clipped rows when flexbox shrank it.
- Regression test `apps/desktop/src/components/chat/composer-dock.test.ts:12` pins the `shrink-0` contract.

No functional issues found.

teknium1:
Thanks @sylvesterkaczmarek for this contribution. This fix path is now on main via #104883 (https://github.com/NousResearch/hermes-agent/pull/104883), merge commit `e412727e7254ff804874e8e04ae97db27d8c4183`.

Exact shrink-0 production patch preserved; behavioral browser tests replace class-string assertion.

Closing this alternative as superseded by the verified merged implementation, not rejecting the reported problem. See the replacement PR for exact live before/after evidence and remaining fidelity limits. Other unresolved issue variants remain open.

Issue #68990, opened 2026-07-21
Thai combining marks dropped/doubled in TUI during streamed rendering (stored content correct)


chongdlh-pixel:
# Thai combining marks are dropped/doubled in the TUI during streamed rendering (stored content is correct)

## Summary

When the agent streams a Thai reply, the terminal **display** corrupts Thai
combining marks — tone marks and above/below vowel signs are dropped, and some
base consonants are doubled. The underlying data is fine: the SSE payload, the
saved session, and copy/paste all contain correct Thai. Only the on-screen
render is wrong. This makes the TUI hard to use for Thai-primary work, since
essentially every streamed reply is visually garbled.

## Environment

- Hermes Agent **v0.19.0** (2026.7.20), commit `11ae6bf0e`, branch `main`
- Node **v26.0.0** (TUI runs under Node, not Bun — confirmed `bun` absent, so the
  JS `stringWidthJavaScript` path is used, not `Bun.stringWidth`)
- `ink` `^6.8.0`, `@hermes/ink` (bundled `packages/hermes-ink`)
- macOS 26.5.2, Apple Terminal (`Apple_Terminal`)

## What I see

Typing a Thai prompt and letting the reply stream in produces, on screen:

| Correct (in data) | Rendered on screen | Defect |
|---|---|---|
| สร้าง | สราง | tone mark ้ (U+0E49) dropped |
| ข้อมูล | ขอมูล | tone mark ้ dropped |
| แจ้งเตือน | แจงเตือน | tone mark ้ dropped |
| คุณ | คคุณ | base ค doubled |
| ฉัน | ฉััน | vowel sign ั (U+0E31) doubled |
| ตอนนี้ | ตอนนนี้ | base น doubled |

The pattern is specifically around Thai marks that are **zero terminal columns
wide**: the base consonant they attach to sometimes doubles, or the mark itself
is dropped, as the streamed text is redrawn incrementally.

## What I ruled out (the data is clean)

This is a **render-only** defect. Verified four ways on the same content:

1. **Model output** — querying the backend directly returns correct Thai; the
   model even self-corrects misspellings, so it is not producing garbled text.
2. **Non-streamed HTTP** — a `stream:false` completion returns correct Thai.
3. **Streamed SSE** — reassembling the `delta.content` chunks yields correct
   Thai. Chunk boundaries split mid-word, but each combining mark stays attached
   to its base within a chunk, so the concatenation is correct.
[...]

PR #98043, opened 2026-08-29
Fix: preserve multi-byte UTF-8 (Thai/Lao/Devanagari/Arabic) across SSE chunk splits

Potz-Aris:
## Problem

When streaming Thai/Lao/Devanagari/Arabic/Hebrew content from any Anthropic-API-compatible provider (Claude, MiniMax, GLM, Kimi), combining marks (vowels, tone marks, harakat) are silently dropped or replaced with U+FFFD — both in the on-screen chat and in the persisted session in `state.db`. ASCII-only content is unaffected.

## Root cause

The Anthropic SDK's `SSEDecoder._iter_chunks` (in `anthropic/_streaming.py`) splits the raw SSE byte stream on `\r\n` / `\n\n` first, then decodes each individual line with strict `bytes.decode("utf-8")`. When a multi-byte UTF-8 character (Thai combining marks are 3 bytes, e.g. U+0E48 mai ek = `0xE0 0xB8 0x88`) straddles two network chunks, the second chunk's leading bytes fail to decode and the offending bytes are dropped or U+FFFD-substituted.

This is the same symptom as Hermes TUI issue #68990 (Ink renderer) but a different layer — this affects **all consumers of the Anthropic SDK in Hermes, not just the TUI**, and persists into the database rather than just on-screen rendering.

## Fix

Add a one-line monkey-patch (`agent/thai_streaming_patch.py`) that replaces `SSEDecoder._iter_chunks` with an implementation that buffers raw bytes through `codecs.getincrementaldecoder("utf-8")` before splitting on SSE boundaries, so multi-byte sequences are reassembled correctly across chunk boundaries.

The patch is:
- **Idempotent** — no-op on second import; safe to call multiple times
- **Graceful** — no-ops when the anthropic SDK isn't installed
- **Hooked into the existing lazy import path** in `agent/anthropic_adapter.py` (`_get_anthropic_sdk`) so it activates at the exact moment the SDK is loaded — no separate wiring needed for any other call site
- **Zero overhead for ASCII-only content** (the original `splitlines` path is still used when the buffered decoder reports a clean boundary)

## Verification

```bash
pytest test_thai_streaming_patch.py
```

Tests cover:
1. **Chunk split mid-UTF-8-sequence** — an SSE event with a Thai combining mark split between two byte chunks round-trips with zero `\ufffd`
2. **Repeated delta stream** — 10 Thai deltas emitted one word at a time (worst-case for chunk-split risk) all decode with their combining marks intact

Both tests pass with the patch applied and fail without it (the second chunk's leading bytes raise `UnicodeDecodeError`).

## Affected

- All Hermes users streaming Thai/Lao/Devanagari/Arabic/Hebrew content
- No-op for users who only stream ASCII / Latin-script / CJK content

## Related

- [Hermes TUI #68990](https://github.com/NousResearch/hermes-agent/issues/68990) — same symptom, different layer (Ink renderer for TUI; this fix addresses the underlying SDK streaming layer that affects all Hermes surfaces including Desktop)

### DIFF
--- a/agent/anthropic_adapter.py
+++ b/agent/anthropic_adapter.py
@@ -67,6 +67,19 @@ def _get_anthropic_sdk():
             pass
         try:
             import anthropic as _sdk
+            # Apply multi-byte UTF-8 streaming fix for Thai/Lao/Devanagari/
+            # Arabic/Hebrew combining marks that the SDK's strict
+            # splitlines-then-decode approach corrupts when SSE chunk
+            # boundaries land mid-multi-byte-sequence. The patch is
+            # idempotent and no-op when not applicable. See
+            # ``agent.thai_streaming_patch`` for the full rationale and
+            # the upstream bug references.
+            try:
+                from agent import thai_streaming_patch as _tsp
+                _tsp.apply_patch()
+            except ImportError:
+                # Patch module missing — degrade gracefully.
+                pass
             _anthropic_sdk = _sdk
         except ImportError:
             _anthropic_sdk = None
--- a/agent/thai_streaming_patch.py
+++ b/agent/thai_streaming_patch.py
@@ -1,0 +1,129 @@
+"""Monkey-patch anthropic SDK streaming decoder to handle multi-byte UTF-8
+split across SSE chunks correctly.
+
+# Problem (issue: Thai/Lao/Devanagari/Arabic combining marks dropped in
+# streaming responses)
+
+The anthropic SDK's ``SSEDecoder._iter_chunks`` (anthropic/_streaming.py)
+splits the raw SSE byte stream on ``\\r\\n`` / ``\\n\\n`` first, then decodes
+each individual line with ``bytes.decode("utf-8")`` (strict mode).
+
+For scripts that contain a mix of single-byte ASCII and multi-byte characters
+whose UTF-8 encoding spans 2-4 bytes (e.g. Thai combining marks like U+0E48
+mai tho = 3 UTF-8 bytes ``0xE0 0xB8 0x88``), the SSE boundary often lands in
+the middle of a multi-byte sequence when ``splitlines()`` is applied before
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** Fixes Thai/Lao/Devanagari/Arabic combining marks dropped when SSE chunk boundaries land mid-UTF-8 sequence.

**What it does**
- `agent/anthropic_adapter.py:16` calls `agent.thai_streaming_patch.apply_patch()` on SDK import; `agent/thai_streaming_patch.py:78` replaces `SSEDecoder._iter_chunks` with a byte-buffered version that yields only on full SSE message boundaries (`\r\r`/`\n\n`/`\r\n\r\n`) so multi-byte sequences straddling chunks are reassembled before downstream decode.
- Patch is idempotent via `_hermes_thai_streaming_patch_applied` flag (`agent/thai_streaming_patch.py:147`), degrades to `errors="replace"` on bad UTF-8 (`agent/thai_streaming_patch.py:103`).
- Repro script `test_thai_streaming_patch.py:202` splits inside a 3-byte Thai char and asserts no `U+FFFD`.

**Non-blocking notes**
- The incremental `codecs.getincrementaldecoder` is instantiated but its output is discarded — buffering is actually `pending += chunk` (raw bytes). Splitting raw bytes on ASCII delimiters (`\n`/`\r` never appear inside valid UTF-8 continuation bytes), so the fix still works; the decoder is only used as a validator. A comment clarifying that would avoid future confusion.
- `test_thai_streaming_patch.py` lives at repo root, not `tests/` — it won't run in CI's default `pytest` discovery. Consider moving to `tests/agent/test_thai_streaming_patch.py` or renaming to `tests/test_thai_streaming_patch.py`.
- `anthropic_adapter.py:19` silently ignores `ImportError` from the patch module — intentional graceful degrade, but a `logger.debug` on that path would help diagnose a missing file.

No functional issues found; root-level test location is the only housekeeping item.

PR #87028, opened 2026-08-15
fix(memory): honor external prefetch timeout

richardclawbot:
## Why
`memory.external_prefetch_timeout` existed in profile config but was never passed to the `MemoryManager`, so every external-memory provider used the hard-coded 8s join.

## Change
- Declare the setting in `DEFAULT_CONFIG`.
- Parse a positive numeric value defensively and wire it into the external `MemoryManager` construction.
- Preserve the 8s default for absent, invalid, and non-positive values.

## Verification
- `uv run --extra dev pytest tests/agent/test_agent_init_memory_prefetch_timeout.py tests/agent/test_memory_provider.py -q` — 68 passed
- `uv run ruff check agent/agent_init.py hermes_cli/config_defaults.py tests/agent/test_agent_init_memory_prefetch_timeout.py`
- `HERMES_HOME=... uv run hermes config check` — config recognizes the setting
- Fresh-profile construction proof: configured 0.5s → manager 0.5s; Terra default/high read back.

## Risk
Only external memory prefetch joins are bounded; recall that exceeds the configured bound is skipped for that turn, exactly as the existing timeout path specifies.

PR #98045, opened 2026-08-29
feat(memory): make external provider prefetch timeout configurable

Navlem:
## Problem

`agent/memory_manager.py` hardcodes `_EXTERNAL_PREFETCH_TIMEOUT_S = 8.0`, but the correct value depends on the external memory provider's recall latency — something the core cannot know.

This is not only a "memory arrives late" issue. In `MemoryManager._prefetch_external` (`memory_manager.py:611-632`) a timed-out worker thread is **not** cancelled:

```python
thread.join(self._external_prefetch_timeout)
if thread.is_alive():
    logger.warning("Memory provider '%s' prefetch timed out after %.1fs; ...")
    return ""
```

The thread keeps running, and lines 611-619 skip prefetch on any later turn while it is still alive. So when a provider's recall reliably exceeds the timeout, prefetch is **never applied at all** — not merely delayed. The only symptom is a repeating warning, which reads like noise rather than "your memory provider is switched off."

Observed with the Hindsight provider against a bank whose local CPU cross-encoder reranked 300 candidates per recall: recall measured 15-22s against the 8s ceiling, 135 warnings accumulated, and prefetch was effectively disabled for the entire period. Raising the timeout restored it immediately.

## Change

Adds `memory.external_prefetch_timeout` (default `8.0`, so existing behaviour is unchanged) and threads it from `agent_init` into `MemoryManager`.

Deliberate choices:

- **`config.yaml`, not an env var** — this is a behavioural setting, not a credential.
- **Invalid values fall back, they do not raise.** A non-numeric or non-positive value in user config must not break startup, so it degrades to the existing default. `MemoryManager`'s own positive-value guard is left intact for direct callers.

## Tests

`tests/agent/test_memory_prefetch_timeout_config.py`:

- Resolution chain: configured float / int / numeric string, absent key, `None`, zero, negative, garbage, empty config.
- Asserts the documented default in `config_defaults.py` matches `_EXTERNAL_PREFETCH_TIMEOUT_S`, so the two cannot silently drift.
- Guards the real `agent_init` call site. The parametrised cases mirror `agent_init`'s resolution logic, which means they would stay green if `agent_init` reverted to a bare `MemoryManager()` — so that case is asserted separately against the source.

Mutation-verified: breaking the wiring fails the guard test (exit 1); restoring it passes (exit 0). 88 tests pass alongside the existing `tests/agent/test_memory_provider.py` suite.

## Verification

Exercised on a live install rather than only in tests: set to `15`, restarted the gateway, confirmed the value resolves through `load_config()` → `MemoryManager` as `15.0`, and confirmed real `hindsight_recall` calls now complete instead of being discarded. Zero timeout warnings from sessions created after the restart.

One behaviour worth noting for reviewers: the timeout is read when a session's agent is constructed, so already-running long-lived sessions keep their previous value until they end. That is consistent with how other `memory.*` settings behave.

### DIFF
--- a/agent/agent_init.py
+++ b/agent/agent_init.py
@@ -1902,7 +1902,17 @@ def init_agent(
             if _mem_provider_name and _mem_provider_name.strip():
                 from agent.memory_manager import MemoryManager as _MemoryManager
                 from plugins.memory import load_memory_provider as _load_mem
-                agent._memory_manager = _MemoryManager()
+                # Prefetch timeout is provider-latency dependent, hence
+                # configurable. A junk value must not break startup: fall back
+                # to MemoryManager's own default.
+                try:
+                    _raw = (mem_config or {}).get("external_prefetch_timeout")
+                    _prefetch_timeout = float(_raw) if _raw and float(_raw) > 0 else None
+                except (TypeError, ValueError):
+                    _prefetch_timeout = None
+                agent._memory_manager = _MemoryManager(
+                    external_prefetch_timeout=_prefetch_timeout
+                )
                 _mp = _load_mem(_mem_provider_name)
                 if _mp and _mp.is_available():
                     agent._memory_manager.add_provider(_mp)
--- a/hermes_cli/config_defaults.py
+++ b/hermes_cli/config_defaults.py
@@ -2040,6 +2040,14 @@ DEFAULT_CONFIG = {
         # "hindsight", "holographic", "retaindb", "byterover".
         # Only ONE external provider is allowed at a time.
         "provider": "",
+        # Seconds a turn waits for an EXTERNAL provider's prefetch (recall)
+        # before continuing without injected memory; the built-in local
+        # provider is unaffected. A timed-out worker is NOT cancelled and
+        # later turns skip while it lives, so a provider slower than this
+        # never gets applied at all. Measure your provider's real p95 and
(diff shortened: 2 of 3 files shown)


alt-glitch:

Duplicate of #87028. Both implement the same `memory.external_prefetch_timeout` config-to-MemoryManager handoff; #87028 is the earlier open and more complete implementation.

rodrigogs:
I've posted the field evidence (Bedrock-only Docker gateway with a local Ollama embedder: 8s timeouts coinciding with 18.4-47.1s embeds and a 400 after 60s, 101/197 agent starts with no injected recall) on #87028, since the triage bot already flagged this PR as a duplicate of it and #87028 is the earlier, more complete head. Two of the same gaps apply here: `cli-config.yaml.example`'s `memory:` block (ends at line 908) doesn't get the new `external_prefetch_timeout` key, and `test_agent_init_actually_passes_the_timeout` asserts on `inspect.getsource(agent_init)` text, which is brittle against the `_init_memory` refactor already on main (agent_init.py:1222-1290) - a call-based test (invoke `_init_memory` directly with a patched `load_memory_provider` and assert on `agent._memory_manager._external_prefetch_timeout`) would survive that. Not duplicating the full note here - see #87028 for the rest.

PR #96842, opened 2026-08-28
feat(desktop): configurable group chat round cap + model-aware limits + token budget guard

beplee:
## Summary

- Replace hardcoded `GROUP_CHAT_MAX_ROUNDS` constant with config.yaml-driven `group_chat.max_rounds` (default 3, range 1-20)
- Add model-aware cap: free models (`:free` suffix or `free_models` allowlist) use `group_chat.max_rounds_free` (null = unlimited, clamped to `HARD_CAP=20`)
- Add token-budget guard: cumulative room-log tokens exceeding `group_chat.token_budget` (default 80k chars) force `exitKind='capped'` independent of round count — the real safety net for pathological cascades
- Add `getGroupChatConfig()`, `isFreeModel()`, `getEffectiveRoundCap()`, `getCurrentModelName()` helpers
- Add config.yaml schema in `cli-config.yaml.example` under `group_chat:`

## Why

Hardcoded 3-round cap is too tight for free-model rooms (where API cost is zero) and too loose for paid-model rooms (where each round burns real money). A single config block drives everything automatically.

## Config

```yaml
group_chat:
  max_rounds: 3              # paid models
  max_rounds_free: null      # null = unlimited (HARD_CAP=20 backstop)
  token_budget: 80000        # hard ceiling per drive (chars, ~20k tokens)
  free_models:
    - upstage/solar-pro4
    - meituan/longcat-2.0
    - tencent/hy3
  hard_cap: 20               # absolute ceiling even for free models
```

## Safety

- Token budget is the real backstop — fires regardless of model type or round count
- `HARD_CAP=20` prevents unbounded free-model cascades (20 rounds × 6 bots = 120 messages max)
- Paid models always capped at `max_rounds` (default 3)
- `exitKind='capped'` reason logged for every termination path

## Test plan

- [ ] Paid model → 3-round cap fires
- [ ] Free model (allowlist) → 20-round cap fires
- [ ] Token budget exhaustion → mid-drive cap fires
- [ ] Model switch mid-drive → cap re-evaluates
- [ ] 6-bot free room → halts cleanly at token budget
- [ ] Empty `free_models` + `:free` suffix → allowlist is sole gate (no bypass)

Refs: #96553

PR #92213, opened 2026-08-22
feat(bots): make group chat round, message, member and history limits per room

oliverhees:
## The problem

A group chat is capped at `GROUP_CHAT_MAX_ROUNDS = 3` and
`GROUP_CHAT_MAX_MESSAGES = 10` per user send, with `GROUP_CHAT_MAX_MEMBERS = 6`
and `GROUP_CHAT_HISTORY_LIMIT = 24`. All four are module constants.

The message cap binds first, and it binds harder than the numbers suggest,
because the two multiply: with N members one round costs N messages. Running
the current loop shows what a room actually gets:

| members | rounds reached | messages | stopped by |
| --- | --- | --- | --- |
| 6 | 2 | 10 | messages |
| 4 | 3 (last one partial) | 10 | messages |

At six members the round cap is unreachable. At four it is reached only
nominally: eight messages are gone after two rounds, so the third round seats
two members and the other two never speak.

That last line is #89545, closed as fixed by #91084. The stranded-harvest fix
addresses the symptom, that late replies were lost and the room looked
abandoned. It does not change the arithmetic: an
Orchestrator → Scout → Builder → Verifier room spends one full round per pass,
and a verifier BLOCK plus a correction handoff plus the re-build is three
passes minimum. Ten messages does not fit that shape, whatever the harvest
does afterwards.

## What this changes

Each of the four axes becomes a per-room override with three states: a number,
off, or absent. Absent inherits the shipped default, and the stored form omits
every axis left alone, so a room that never opens the settings stores nothing
and behaves exactly as it does today. Defaults are unchanged.

Turning rounds or messages off hands control to a safety brake, defaulting to
50 rounds / 200 messages, which is itself switchable.

## On unbounded rooms, given #91481

#91481 is open: two Telegram bots exchanged 132 messages with no loop guard
anywhere in the admission path. Offering an unlimited group room in that
climate deserves an explicit answer rather than silence, so:

- The brake is **on by default** whenever an axis is switched off. Reaching
  "no limit at all" takes two deliberate steps on two separate controls, and
  the editor states in plain words what that means before the user gets there.
- Unlike the Telegram path, this loop already has two exits that do not depend
  on a counter. `spokeThisRound === 0` ends the drive when every member passes
  a round, and a new user message bumps the room epoch, which the loop honors
  at the next member boundary. An unlimited room is a room that stops counting,
  not a room you cannot get out of.
- A room that stops on its brake records a `safety` activity event naming the
[...]

PR #98047, opened 2026-08-29
feat(desktop): drive Bot Mode room limits from config.yaml

### LINKED ISSUE #98004: [Feature]: Per-group auto-continue budget for Bot Mode rooms (item 4 of #89545)
## Problem or Use Case

#89545 landed the honest-stop half of the safety-cap story: a room that hits `GROUP_CHAT_MAX_ROUNDS` / `GROUP_CHAT_MAX_MESSAGES` now reports `turn stopped at the round/message cap` instead of masquerading as consensus. That was the right first move and it works.

The fourth item of that issue's own "smallest useful fix" list never shipped:

> Optionally allow an explicit per-group auto-continue budget, still bounded and off by default.

Without it, the honest stop is still a full stop. Every capped drive needs a human to type `continue`, which is exactly the manual babysitting the visible-stop work was meant to make unnecessary.

Concretely, in a six-member LifeOS room (coordinator, three domain coordinators, a Notion data-plane operator, a silent reliability bot), a routine month-close runs:

```
user request
  -> coordinator triages
  -> finance coordinator scopes the write
  -> notion-expert reads, hits an upstream 502 on a rate source
  -> finance coordinator picks the fallback source
  -> notion-expert writes and verifies
```

That is five to six hops of genuinely sequential work. Three rounds cannot express it, so the room stops mid-workflow with a bot holding an addressed handoff, and the user resumes it by hand — often several times for one task.

Two aggravating factors are worth naming, because they make the cap bind sooner than the numbers suggest:

- `resolveGroupResponders` (`group-rounds.ts:139-140`) returns every member when no one is `@`-mentioned, so an unaddressed user message spends six of the ten messages in round one.
- The responder set only grows across rounds: each `@`-mention adds a permanent speaker, and idle members still pay a model call to answer `(pass)`.

## Proposed Solution

Add the bounded, default-off budget #89545 already specified, configured per group rather than hardcoded:

```yaml
bot_mode:
  group_chat:
    auto_continue_budget: 0   # default 0 = today's behavior, unchanged
```

When a drive exits `capped` (never `settled`) and the last member message contains a valid `@bot` handoff, spend one unit of budget and start another bounded drive on the same thread, preserving per-member watermarks. Decrement per continuation, stop at zero, and surface the remaining budget in the same activity line that reports the cap today. A user send still supersedes a running drive.

The hard caps stay exactly as they are — this bounds *how many* bounded drives a single user send may chain, which is the shape #89545 asked for. Runaway control is preserved: the budget is finite, explicit, and off unless configured.

Two smaller changes would reduce how often the budget is needed at all, and could ship separately:

- Let a member with nothing to add skip without a model round-trip. Today `(pass)` costs a full call (~4s here), so idle members burn both wall-clock and the message budget.
- Reconsider the grow-only responder set, or let a room opt into "mentioned members only" after the first round.

## Alternatives Considered

**Raising the constants locally.** Works, but `apps/desktop/src/plugins/hermes-bots/group-chat.ts:1208-1212` are plain `export const` values imported statically in eight places in `group-rounds.ts`, so there is no override path: not config, not a runtime plugin (`PluginContext` exposes `register`/`rest`/`socket`/`os`/`storage`/`i18n` — nothing that can rebind another plugin's module binding), not an env var (and `AGENTS.md` rules those out for behavioral config anyway). A local edit is also parked into `git stash` by `hermes update --keep-stash` on every update, so it silently reverts. The comment directly above those constants already flags that "deciding the shape of the override belongs to those PRs".

**Prompt-level discipline.** Real gains, and worth doing regardless — routing in one hop instead of relaying through a coordinator, and instructing off-domain members to pass immediately, roughly halved our round consumption. It does not close the gap for genuinely multi-hop work, and it cannot help a workflow whose hop count is irreducible.

**Just typing `continue`.** What we do today. It works and loses no state, but it makes a bounded, well-understood workflow depend on a human being present, which is the thing #89545 set out to fix.

## Feature Type

Configuration option

## Scope

## Contribution

I'd like to implement this myself and submit a PR.

## Environment

Hermes Agent v0.20.6, macOS Desktop, six local bots in one group.

andredezzy:
## What does this PR do?

A Bot Mode room stops after 3 serial rounds and 10 member messages. That suits a quick two-bot exchange and cuts short a workflow whose hops are genuinely serial: a coordinator triaging to a domain expert, to a data operator, and back needs more passes than three. The room then reports `turn stopped at the round/message cap` with a bot still holding an addressed handoff, and the user resumes it by hand.

This reads the ceilings from a `group_chat` block in `config.yaml`, falling back to today's values when it is absent, so the person paying for the model calls decides what a room may spend.

Config is advisory, never authoritative. Each value is clamped to a hard ceiling (20 rounds / 100 messages / 20 continuations), and anything below 1 or unparseable reads as *unset* rather than as zero — a typo can neither silence a room nor uncap it. The caps themselves are unchanged for anyone who does not configure them.

This is a port of the design in #96842 (@beplee), which was closed unmerged when #96726 deleted the `plugin.js` it patched. The closure note asked for exactly this:

> **What comes next:** port the configurable round cap ... to the new module surface after #96726 merges.

Two deliberate departures from the original:

- **No new gateway method.** The original called `config.get` with `key: 'group_chat'`, but that handler rejects unknown keys (`unknown config key: group_chat`, `methods_config.py:558`) and the call sat inside a `try/catch` that fell back to defaults — so the config would never have been read. This uses the existing `full` key instead, and touches no Python.
- **Scope.** Model-aware caps and the token-budget guard from #96842 are left out; this is the round/message/continuation seam only. #92213 (per-room limits) can layer on the same resolver.

## Related Issue

Fixes #98004

## Type of Change

## Changes Made

- `apps/desktop/src/plugins/hermes-bots/group-chat.ts` — `resolveGroupChatLimits()` (pure, clamped, junk-tolerant), the hard-cap constants, and `refreshGroupChatLimits()` reading `config.get` `full`.
- `apps/desktop/src/plugins/hermes-bots/group-rounds.ts` — `runGroupChatRounds` snapshots the limits once per drive instead of importing the constants.
- `apps/desktop/src/plugins/hermes-bots/plugin.tsx` — refresh once at plugin load.
- `cli-config.yaml.example` — document the `group_chat` block.
- `group-rounds.test.ts` — 6 tests; `plugin-panes.test.tsx` — mock the new export.

Resolution happens once at load and is snapshotted per drive rather than read inside the round loop, which is hot — a gateway round-trip there would be a stall the user feels. A config edit reaches a room on its next send. Reading fails silently to the shipped defaults: a room that runs beats a room that refuses to start because config could not be read.

## How to Test

1. Without any `group_chat` block, open a Bot Mode room with several members and send an unaddressed message. Behavior is unchanged: the drive stops at 3 rounds / 10 messages.
2. Add to `config.yaml`:
   ```yaml
   group_chat:
     max_rounds: 12
     max_messages: 40
   ```
3. Restart Desktop and repeat. The room now runs up to 12 rounds and 40 messages, and a serial coordinator → expert → operator → back workflow completes instead of stopping mid-handoff.
4. Set `max_rounds: 99999` and confirm it clamps to 20; set `max_rounds: 0` (or `"many"`) and confirm it falls back to 3 rather than muting the room.

Automated:

```
npx vitest run src/plugins/hermes-bots/     # 562 passed (57 files)
npx tsc --noEmit -p tsconfig.json           # clean
npx eslint <changed files>                  # 0 errors, 0 warnings
```

The 8 new tests were written failing first and cover: defaults when config is absent, configured values winning, clamping at the hard cap, junk (`0`, negative, `NaN`, a word, empty/blank string, `null`, `undefined`, `[]`, `{}`, `true`) falling back rather than silencing, quoted numbers being read, fractional values flooring, and an end-to-end drive posting past the default message cap under a raised ceiling while still stopping.

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate — #96842 is the closed original this ports; #92213 is complementary (per-room), not overlapping
- [x] I've run the suites this change can affect — `vitest` over `src/plugins/hermes-bots/` (562 passed), `tsc`, and `eslint`. The diff is Desktop TypeScript only and touches no Python; `pytest tests/ -q` was still running locally when I opened this, so I am not claiming a green run I did not watch finish. Happy to post the result if CI does not cover it.
- [x] I've tested on my platform: macOS 26.6.2, Desktop

- [x] I've updated relevant documentation — `cli-config.yaml.example`
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys
- [x] I've considered cross-platform impact — N/A, no platform-specific code

## Credit

The design is @beplee's from #96842; this ports it onto the post-#96726 modules and drops the gateway dependency that would have kept it from working.


andredezzy:
Following up on the test-suite caveat in the checklist, now that I have the numbers.

At this PR's commit (`10e7dae52`), `tests/tui_gateway/` + `tests/gateway/relay/` run **973 passed, 1 failed**. That single failure is `test_gui_surface_toolsets.py::test_holds_exactly_the_gui_affordances`, and it passes both on clean `main` and on this commit when run on its own (10 passed) — shared state across a batch run, not this change.

Worth flagging for anyone running the suite locally: `pytest-asyncio` is not in the base install, and without it `tests/gateway/relay/test_ws_transport.py` and `test_relay_going_idle.py` fail at collection. With it installed they pass (7 passed).

The diff itself is Desktop TypeScript only and touches no Python, so the Python suites are unaffected by construction — I just wanted to verify rather than assert that.

andredezzy:
Superseded by a clean branch — a main merge pulled unrelated files into this PR (21 files instead of 6). Reopening the same change isolated; closing this to keep review honest.

PR #98055, opened 2026-08-29
fix(tui): recover remote sessions after WebSocket drops

### LINKED ISSUE #98042: TUI: remote client ws drop mid-turn leaves session unrecoverable — sticky 'gateway exited' flag, queued msgs, phantom 'another process is using this session' (backend healthy)
## Summary

When a **remote TUI client** (laptop over LAN to `hermes dashboard --host <ip> --port 9119`) drops its websocket **mid-turn** (laptop sleep, SIGHUP, transient network), the server-side turn continues and completes normally — but the reconnected client never receives the result. The TUI shows a **permanently sticky `─ gateway exited` status-bar flag**, queues subsequent user messages while the original turn is still running, and on later reconnects shows "Another Hermes process is using this session; waiting for it to finish" — even though the "other process" is just the same session's still-running turn.

Net effect: a routine client blink is indistinguishable from a dead backend, and the session appears unrecoverable (`/restart` only restarts the client; the queued state and stale flag persist).

## Distinguishing this from near-duplicates

- **#94935** — completed turn never rendered, `ws_orphan_reap`: local TUI, idle-window case. Related but local + idle-triggered.
- **#97764** — desktop renderer frozen after mid-turn ws drop, RPC rejected "not in memory": desktop app, resume failure. Related; this one is the TUI + remote variant.
- **#39013** — "gateway exited" + "session not found" after rapid messages: stdin-EOF gateway respawn path, different mechanism (gateway actually exits).

Here the **backend stays up throughout** (verified: dashboard and gateway each had 13.5h+ continuous uptime across the incident window). The failure is entirely in the reconnect/recovery UX for remote clients.

## Environment

- Hermes Agent v0.20.5 → updated to v0.20.6 (2026.8.27) mid-investigation; behavior observed on both
- Backend: `hermes dashboard --no-open --host 192.168.50.239 --port 9119`, launchd-supervised (`KeepAlive=true`), macOS 26.3 arm64 (always-on studio machine, ethernet)
- Client: Hermes TUI on a MacBook on the same LAN (192.168.50.187)
- Model: moonshotai/kimi-k3 via nous; delegation model deepseek/deepseek-v4-pro-0813

## Steps to reproduce

1. On machine A (studio): run `hermes dashboard --host <lan-ip> --port 9119` (launchd KeepAlive recommended).
2. On machine B (laptop): open a TUI session against A over LAN.
3. Start a **long turn** (in our case ~800s turns from a heavy subagent session — the long duration is what makes the window easy to hit).
4. Mid-turn: sleep the laptop lid, or otherwise sever B's connection (SIGHUP).
5. Reopen / let the TUI reconnect.

## Expected

- Client reconnects and resumes the in-flight stream (or at minimum, picks up the completed turn result).
- Status bar reflects the true state: backend healthy, client reconnected.
- Queued messages flush once the turn ends.

## Actual

- Banner: `gateway exited — recovering your session (any in-flight reply was lost)` — the "in-flight reply was lost" claim is wrong; the turn completes server-side (`status=complete`, persisted to state.db).
- Status bar flag `─ gateway exited` **never clears** for the life of the TUI process.
- New messages show `queued (n)` while the in-flight turn runs; on reconnect the client gets "Another Hermes process is using this session; waiting for it to finish" (the session's own running turn is the blocker).
- `/restart` in the TUI has no effect — it's client-only; nothing server-side is broken.

## Evidence

**Server-side: backend never died.** Dashboard (pid 92597) and gateway (pid 92852) each had 13.5h continuous uptime spanning the incident. Gateway log shows zero exits/restarts overnight.

**Client-side: repeated SIGHUP + 1006 drops overnight** (`~/.hermes/logs/tui_gateway_crash.log` on the laptop):

```
2026-08-29T02:30:12Z  transport exit code=1006 reason=gateway websocket closed (1006)
2026-08-29T02:30:12Z  graceful-exit received signal=SIGHUP → killing gateway
2026-08-29T02:30:15Z  uncaughtException: Error: write EIO
...13 SIGHUP/1006 events between 02:30 and 04:11 local...
```

**Turn completes normally after the client is gone** (backend `gui.log`):

```
2026-08-29 11:01:47  tui prompt accepted: session_key=202608...4824 kind=user chars=60
2026-08-29 11:15:16  tui turn finished: session_key=202608...4824 status=complete duration=809.2s
```

**Later reconnect attempt is gated by the still-running turn** (backend log):

```
⏳ Another Hermes process is using this session; waiting for it to finish before starting your turn...
```

**`session_turn_leases`** in state.db shows the lease holder is the same session's in-flight turn — the "other process" is a phantom created by the disconnect.

## Suggested investigation

1. **Clear the `gateway exited` flag on successful reconnect** — the transport layer already schedules reconnects (`scheduling gateway reconnect in 1000ms`) and succeeds; the UI state never reflects recovery.
2. **Re-attach to the in-flight turn after reconnect** instead of discarding the local turn view — the server has the full state; the client treats the turn as lost.
3. **The "another process is using this session" gatekeeper message is misleading** when the holder is the same session's own in-flight turn from a disconnected client — it should say "this session has a turn in flight (started HH:MM, N elapsed); waiting…" and ideally offer interrupt-on-attach.

## Impact

For remote setups (laptop → always-on studio machine) with long-running agent turns, any laptop sleep mid-turn produces a session that *looks* permanently dead. Data is never lost, but the UX signal ("gateway exited", queued messages, stuck flag) teaches the user the backend is unreliable when it is in fact healthy — and the only discovered workaround is abandoning the session.

fangliquanflq:
## What does this PR do?

Remote TUI sessions now receive gateway events again after a mid-turn WebSocket reconnect. Without this fix, the replacement socket connects successfully but its `gateway.ready` and subsequent session events remain buffered forever, leaving the UI stuck on `gateway exited`, preventing the recovery resume, and stranding queued messages.

### Symptom

After a remote TUI WebSocket closes mid-turn and reconnects, the status remains `gateway exited`. The completed or continuing turn does not return to the TUI, and later messages remain queued.

### Impact

Remote TUI users can lose access to an otherwise healthy, persisted session after a laptop sleep or transient network drop. The backend turn can continue and complete, but the mounted client never observes the reconnect events needed to reattach and reconcile its state.

### Bug Cause

**Trigger:** `ui-tui/src/gatewayClient.ts:335` / `resetStartupState()` during WebSocket replacement.

**Causal chain:**
1. A subscribed remote TUI loses its WebSocket while a turn is running.
2. The exit handler starts a replacement transport, and `resetStartupState()` resets `subscribed` to false even though the UI event consumer remains mounted.
3. `gateway.ready` and all later session events are buffered, but `drain()` only runs during the original mount, so recovery never resumes the persisted session and the UI remains stale.

**Why it is wrong:** Subscription ownership belongs to the long-lived UI consumer, not to an individual child process or WebSocket generation. Resetting it during transport replacement disconnects the mounted listener from every future event.

**Working sibling / contrast:** The initial attach path works because the mount effect calls `drain()` once after registering the event listener. A reconnect has the same listener but no second mount-time drain.

**Ruled out:** The gateway process and session persistence are not the failure point. The server already keeps the live session attachable and returns persisted history or in-flight state from `session.resume`; the client never receives the replacement socket's `gateway.ready` event that initiates that resume.

### Fix

Preserve the consumer subscription across transport startup resets while still rejecting stale RPCs, invalidating old drain generations, and clearing transport-specific buffers. A regression test closes a subscribed WebSocket, lets the existing exit subscriber replace it, and verifies that `gateway.ready` from the new socket reaches the mounted consumer.

## Related Issue

Closes #98042

## Type of Change

## Changes Made

- `ui-tui/src/gatewayClient.ts` - keep the mounted event subscription active when replacing a failed transport.
- `ui-tui/src/__tests__/gatewayClient.test.ts` - cover event delivery after a subscribed WebSocket reconnect.

## How to Test

1. Run the focused WebSocket reconnect regression suite.
2. Confirm type checking and linting complete without errors.

```bash
cd ui-tui
npm test -- --run src/__tests__/gatewayClient.test.ts
npm run typecheck
npm run lint
```

### Code

- [x] I've run the relevant TUI test suite and all focused tests pass

- [x] Documentation update: N/A - no user-facing configuration or workflow changed
- [x] `cli-config.yaml.example`: N/A - no configuration keys changed
- [x] `CONTRIBUTING.md` or `AGENTS.md`: N/A - no architecture or workflow changed
- [x] Cross-platform impact considered - the transport state transition is platform-independent
- [x] Tool descriptions/schemas: N/A - no tool behavior changed

## Screenshots / Logs

Focused regression result: 19 tests passed in `src/__tests__/gatewayClient.test.ts`.

### DIFF
--- a/tests/tui_gateway/test_resume_live_lazy_session.py
+++ b/tests/tui_gateway/test_resume_live_lazy_session.py
@@ -7,6 +7,7 @@ by stored key or pending title — which hard-404'd "session not found" for
 every bot that had never spoken (community + Teknium repro, Aug 2026).
 
 Contract:
+- resume by runtime id reattaches to the live in-memory record;
 - resume by stored session_key reattaches to the live in-memory record;
 - resume by pending title reattaches likewise;
 - the match is scoped to the SAME profile home — an unscoped resume of a
@@ -50,6 +51,14 @@ def live_lazy_session(home):
 
 def _resume(params):
     return srv._methods["session.resume"](1, params)
+
+
+def test_resume_by_runtime_id_reattaches(live_lazy_session):
+    sid, record = live_lazy_session
+    out = _resume({"profile": "ops", "session_id": sid, "omit_messages": True})
+    assert "error" not in out, out
+    assert out["result"]["session_id"] == sid
+    assert out["result"]["stored_session_id"] == record["session_key"]
 
 
 def test_resume_by_stored_key_reattaches(live_lazy_session):
--- a/tui_gateway/methods_session.py
+++ b/tui_gateway/methods_session.py
@@ -431,8 +431,10 @@ def _(rid, params: dict) -> dict:
                 # the profile, the open/send then resumes it, and this hard 404
                 # ("session not found") killed messaging for exactly the bots
                 # that had never spoken. Match the in-memory registry by stored
-                # key or pending title, scoped to the SAME profile home this
-                # resume targets, and hand the caller the live record.
+                # runtime id, stored key, or pending title, scoped to the SAME
+                # profile home this resume targets, and hand the caller the
+                # live record. Runtime-id lookup is required when a replacement
+                # WebSocket reattaches the session after a mid-turn drop.
                 # (Nested per method_ctx rebinding — module helpers are
                 # invisible from installed handlers.)
                 def _find_live_unpersisted(needle: str, home) -> str:
@@ -443,7 +445,8 @@ def _(rid, params: dict) -> dict:
                         if (record.get("profile_home") or None) != want_home:
                             continue
                         if (
-                            str(record.get("session_key") or "") == needle
+                            live_sid == needle
+                            or str(record.get("session_key") or "") == needle
                             or (record.get("pending_title") or "") == needle
                         ):
                             return live_sid
--- a/ui-tui/src/__tests__/createGatewayEventHandler.test.ts
+++ b/ui-tui/src/__tests__/createGatewayEventHandler.test.ts
@@ -1018,7 +1018,7 @@ describe('createGatewayEventHandler', () => {
 
     onEvent({ payload: {}, type: 'gateway.ready' } as any)
 
-    await vi.waitFor(() => expect(resumeById).toHaveBeenCalledWith('sess-crashed'))
+    await vi.waitFor(() => expect(resumeById).toHaveBeenCalledWith('sess-crashed', true))
     expect(newSession).not.toHaveBeenCalled()
     // One-shot: the ref is consumed so a later ordinary restart forges/resumes
     // per config instead of re-resuming the recovered session.
--- a/ui-tui/src/__tests__/gatewayClient.test.ts
+++ b/ui-tui/src/__tests__/gatewayClient.test.ts
@@ -734,6 +734,89 @@ describe('GatewayClient websocket attach mode', () => {
     }
   })
 
+  it('releases sibling session events after focused session replay', async () => {
+    process.env.HERMES_TUI_GATEWAY_URL = 'ws://gateway.test/api/ws?token=abc'
+    const gw = new GatewayClient()
+    const events: Array<{ seq?: number; session_id?: string; type: string }> = []
+
+    try {
+      gw.on('event', event => events.push(event as { seq?: number; session_id?: string; type: string }))
+      gw.on('exit', () => gw.start())
+      gw.start()
+
+      const first = FakeWebSocket.instances[0]!
+
+      first.open()
+      gw.drain()
+      await Promise.resolve()
+      first.message(
+        JSON.stringify({
+          jsonrpc: '2.0',
+          method: 'event',
+          params: { payload: { replay_epoch: 'epoch-1' }, type: 'gateway.ready' }
+        })
+      )
+
+      for (const sessionId of ['s1', 's2']) {
+        first.message(
+          JSON.stringify({
+            jsonrpc: '2.0',
+            method: 'event',
+            params: { payload: {}, seq: 1, session_id: sessionId, type: 'message.start' }
+          })
+        )
+      }
+
+      first.close(1006)
+
+      const second = FakeWebSocket.instances[1]!
+
+      second.open()
+      second.message(
+        JSON.stringify({
+          jsonrpc: '2.0',
+          method: 'event',
+          params: { payload: { replay_epoch: 'epoch-1' }, type: 'gateway.ready' }
+        })
+      )
+
+      const replay = gw.replaySessionEvents('s1')
+
+      await vi.waitFor(() => expect(second.sent.some(raw => JSON.parse(raw).method === 'session.events.since')).toBe(true))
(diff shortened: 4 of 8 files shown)


Enough1122:

**Overall:** Fixes remote TUI stalling on `gateway exited` after a mid-turn WebSocket reconnect.

**What it does**
- `ui-tui/src/gatewayClient.ts:588` preserves `subscribed` across WebSocket replacement (previously cleared in `onTransportExit` and never drained), adds `lastSeenSeq`/`replayEpoch`/`replayHold` tracking (`ui-tui/src/gatewayClient.ts:549`), sequence-deduped `publishIfNewer` (`ui-tui/src/gatewayClient.ts:615`), `hasReplayWatermark`/`replaySessionEvents`/`finishReconnectRecovery`/`flushReplayHold` (`ui-tui/src/gatewayClient.ts:632`).
- `tui_gateway/methods_session.py:51` extends `_find_live_unpersisted` to match `live_sid == needle` so resume by runtime id reattaches (`tests/tui_gateway/test_resume_live_lazy_session.py:17`).
- `ui-tui/src/app/useSessionLifecycle.ts:432` threads `recovering` flag through `resumeById(id, recovering?)`, uses `omit_messages` + `session.events.since` replay with `reconcile` fallback; `ui-tui/src/app/createGatewayEventHandler.ts:369` calls `resumeById(recoverSid, true)`.

**Non-blocking notes**
- New state (`replayHold: Map<string, GatewayEvent[]>`) holds parked events until `finishReconnectRecovery` — bounded (only while a reconnect is pending) but worth noting in a long-lived TUI with many sibling sessions the hold could briefly buffer `MAX_BUFFERED_EVENTS` across sessions; `flushReplayHold` drains eagerly.
- `REPLAY_REQUEST_TIMEOUT_MS = 10_000` (`ui-tui/src/gatewayClient.ts:540`) is shorter than `REQUEST_TIMEOUT_MS` — good, avoids long stall on `session.events.since`.
- Epoch comparison uses `replay_epoch` from `gateway.ready` payload — ensures a backend restart (different seq namespace) forces full resume rather than replay.

No functional issues found. New tests in `ui-tui/src/__tests__/gatewayClient.test.ts:76` (4 cases) cover reconnect, detached terminal replay, sibling release, and no-watermark path.

fangliquanflq:
Thanks for the review. I verified each note against the current head and the base behavior:

- The reconnect hold is bounded to sessions with existing sequence watermarks and is released by `replaySessionEvents()` or `finishReconnectRecovery()` on successful, fallback, and no-watermark recovery paths; the sibling-session tests cover the eager release.
- The 10-second replay timeout is intentional: failure returns `reconcile: true`, so the lifecycle falls back to an authoritative full `session.resume` instead of waiting for the normal RPC timeout.
- A changed `replay_epoch` clears stale watermarks/holds and forces that same full-resume path, preventing sequence values from different backend processes from being compared.

No code change is needed for these informational points. I ran the two focused TUI suites (127 tests passed), TypeScript typecheck, and lint (0 errors; 2 pre-existing hook warnings). The current head also has all required GitHub checks passing, including Python tests.

PR #98058, opened 2026-08-29
fix(agent): use canonical EXCLUDED_SKILL_DIRS in skill scan and exclude _archive

### LINKED ISSUE #98044: Skills scan: inline exclusion set in skill_commands.py drifted from EXCLUDED_SKILL_DIRS; non-dot archive dirs (_archive) are scanned and archived skills stay live

Parker-Fawcett:
Fixes #98044

**What & why**
- `scan_skill_commands()` hardcoded `{'.git','.github','.hub','.archive'}` while `skill_utils.EXCLUDED_SKILL_DIRS` has 14 entries. Skills under `venv/`, `node_modules/`, etc. were incorrectly scanned from one code path but not the other.
- Non-dot archive dirs (`_archive`) — a natural Finder-visible alternative to hidden `.archive` — were not excluded anywhere, so archived skills remained invocable.

**Fix**
- Import the authoritative set instead of inlining a subset
- Add `_archive` to the canonical exclusion set

Verified with headless repro: skills under `_archive`, `venv`, `node_modules`, etc. are now correctly skipped while normal skills remain discoverable.

### DIFF
--- a/agent/skill_commands.py
+++ b/agent/skill_commands.py
@@ -14,6 +14,7 @@ from typing import Any, Dict, Optional
 
 from hermes_constants import display_hermes_home
 from agent.prompt_cache_boundary import register_stable_prefix
+from agent.skill_utils import EXCLUDED_SKILL_DIRS
 from agent.skill_preprocessing import (
     expand_inline_shell as _expand_inline_shell,
     load_skills_config as _load_skills_config,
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #98054 — both implement the same _archive exclusion and canonical skill-scan policy. #98054 was opened first and includes focused regression coverage.

PR #21427, opened 2026-05-07
Restore session cwd on resume

gogadoro0524:
Hi! This is a follow-up to #19154 with the missing consumer added.

In the previous PR, the schema/storage plumbing for session workspace context was clean, but it only wrote `workspace_path` / `last_cwd` and nothing read them. This version keeps the same narrow scope, but wires the stored cwd into core resume behavior so the schema bump has an immediate reader.

## Summary

- Adds `workspace_path` and `last_cwd` columns to the session store.
- Stores the current `TERMINAL_CWD` / process cwd when CLI and agent sessions are created.
- Updates `last_cwd` after terminal tool calls by reading the active terminal environment cwd.
- Restores `last_cwd` into `TERMINAL_CWD` during resume flows:
  - early `--resume` preload
  - `_init_agent()` fallback resume
  - interactive `/resume <session>`
- Keeps local restores conservative: missing/non-absolute local paths are ignored instead of breaking resume.

This means dashboards and remote clients can now rely on persisted workspace context, while the CLI itself also consumes that metadata when resuming a session.

## Tests

- `python3 -m py_compile session_workspace.py hermes_state.py cli.py run_agent.py`
- `.venv/bin/python -m pytest tests/test_hermes_state.py tests/cli/test_resume_display.py -q`
  - `254 passed`

I also attempted the broader local suite with the canonical pytest settings. The run reached `20212 passed`, but this local environment is missing optional/full extras used by unrelated tests (`acp`, `fastapi`, `numpy`, etc.), so that full run was not green locally.

PR #17839, opened 2026-04-30
feat: add workspace support to API Server endpoints

zxd-666:
## Summary

Add `X-Hermes-Workspace` request header support to `/v1/chat/completions` and `/v1/responses` API endpoints, enabling API clients (WebUI, external tools) to control the agent's working directory.

## Motivation

Currently, the agent's working directory is determined by `TERMINAL_CWD` environment variable, but API clients have no way to set it. This forces WebUI implementations to either:
- Directly `import AIAgent` (Python-only option, not available for Node.js backends like hermes-web-ui)
- Accept the default workspace for all sessions

## Changes

**Single file: `gateway/platforms/api_server.py`**

- **`_validate_workspace(path)`** — New static method that validates workspace paths:
  - Resolves symlinks via `os.path.realpath()`
  - Blocks path traversal attempts (`..` components)
  - Restricts to user's home directory
  - Verifies path existence

- **`_create_agent()`** — Extended to accept optional `workspace` parameter:
  - Temporarily sets `os.environ["TERMINAL_CWD"]` before agent creation
  - Restores original value after creation (thread-safe)

- **`_run_agent()`** — Extended to accept and pass through `workspace` parameter

- **`_handle_chat_completions()`** — Parses `X-Hermes-Workspace` header, validates, and passes to agent pipeline. Echoes active workspace in SSE response headers.

- **`_handle_responses()`** — Same workspace support for the Responses API endpoint.

## Usage

```bash
curl -X POST http://localhost:8642/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "X-Hermes-Session-Id: my-session" \
  -H "X-Hermes-Workspace: /Users/me/my-project" \
  -d '{"messages": [{"role": "user", "content": "read README.md"}]}'
```

## Testing

All 186 existing tests pass (119 api_server + 67 related). No test changes required — the new parameter is optional and defaults to current behavior.

## Backwards Compatibility

Fully backwards compatible. When `X-Hermes-Workspace` is not provided, behavior is identical to before.

Issue #29531, opened 2026-05-20
Per-session working directory for gateway (OpenAI-compatible API) sessions


pmos69:
## Problem

Hermes' working directory (`terminal.cwd` → `TERMINAL_CWD`) is process-global. The gateway serves many concurrent sessions — the OpenAI-compatible `/v1/chat/completions` endpoint, messaging platforms — from a single process, all sharing that one cwd. An individual API-driven session has no way to operate in its own working directory: the `file`, `terminal`, and `code_execution` tools all resolve against the single global value.

This makes per-project or per-conversation workflows impossible for any client of the OpenAI-compatible API — every concurrent session is pinned to the same directory.

## Prior art that doesn't close this gap

- **`terminal.cwd` / `TERMINAL_CWD`** — global, fixed at gateway startup.
- **`/set-workspace` (#23535)** — registered `cli_only=True`; runs in the interactive CLI REPL and mutates the *global* env var, so it neither reaches API sessions nor isolates concurrent ones.
- **ACP sessions** — these *do* get a per-session cwd, via `register_task_env_overrides` + per-`task_id` path resolution in `file_tools` (`_resolve_path_for_task`). That path is exclusive to the ACP adapter.
- **Per-session cwd storage (#19292, #21427)** — records/restores a session's cwd for identification and `--resume`, not for concurrent isolation.
- **Gateway cwd fixes (#29365, #27488)** — make gateway sessions honor the *configured* global cwd; still a single value.

## Observation

The per-session-cwd machinery already exists — for ACP. `register_task_env_overrides` plus per-`task_id` resolution gives an ACP session its own working directory while its tools run. It is simply not exposed to OpenAI-compatible API sessions.

## Proposal

Allow an OpenAI-compatible API session to specify a working directory, routed through the same per-task override mechanism the ACP adapter already uses, so that:

- the `file`, `terminal`, and `code_execution` tools resolve against that session's directory;
- ideally, project context-file discovery (`AGENTS.md` / `.hermes.md` / `.cursorrules`) follows the same directory (cf. #11515, the analogous ACP gap);
- concurrent sessions with different working directories do not interfere.

## Open questions for maintainers

- **Transport** — a field on the chat-completions request body? a session-creation parameter? a gateway-routable `/set-workspace`?
- Should context-file discovery follow the per-session cwd, or stay on the global value?
- How should this compose with the session-cwd storage in #19292 / #21427?

Happy to implement this once there's a steer on the preferred transport.

[...]

PR #98060, opened 2026-08-29
fix(gateway): persist per-session cwd for messaging-gateway sessions

narigondelsiglo:
## What does this PR do?

Messaging-gateway chats (Telegram/Discord/Slack/…) never persisted their working directory to `state.db`: only CLI/TUI did. The Projects sidebar groups `sessions` rows by `cwd`/`git_repo_root`, so every gateway chat grouped under the launch dir forever — even after the agent had `cd`-ed into a project. The cwd was already tracked in memory (`record_session_cwd` after every terminal command); it just never reached the database.

This PR dual-writes it: after a terminal command that legitimately reported its cwd (same guards as the existing in-memory record — no transient `workdir` override, no unobserved/interrupted command), gateway sessions persist `cwd`, `git_repo_root`, and `git_branch` to `state.db`, best-effort and never failing the user's command.

Two design decisions, documented in code:

- **Eager per-command** (not session-idle batch): the correct guards already exist inline at the single call site, and there is no intermediate state to lose.
- **Synchronous git probe** (2s timeout, unlike TUI's async enrichment): the command already returned, so the probe is off the hot path; the DB write still goes through `update_session_cwd`'s generation-claim so async enrichment surfaces can never overwrite it.

## Related Issue

Fixes the gateway part of #29531

## Type of Change

## Changes Made

- `hermes_state.py`: new `SessionDB.update_session_cwd_by_session_key()` — gateway rows mint timestamped uuid ids (`20260828_100000_<uuid8>`), so a routing `session_key` cannot reach `update_session_cwd` directly. The helper resolves the key to the row it currently routes to (the still-open row, else the most recently started — reset chains keep several rows per key) and delegates.
- `tools/terminal_tool.py`: `_probe_git_meta()` (sync `git rev-parse` ×2, 2s timeouts, failure yields `(None, None)` which the DB treats as "don't clobber") and `persist_gateway_session_cwd_to_db()` (scope-gated to `agent:`-prefixed keys so CLI/TUI keep their own paths; `try/except → logger.debug`, never raises). Wired into the existing post-command dual-write block.
- `tests/state/test_update_session_cwd_by_session_key.py`: 5 unit tests (open-row preference over older/newer closed rows, closed-row fallback, unknown key, empty-arg rejection).
- `tests/tools/test_terminal_persist_session_cwd.py`: 9 E2E tests driving the real `terminal_tool` against a temp `HERMES_HOME` — happy path with git metadata, workdir-override skip, interrupted-command skip, CLI/TUI scope skips, short `agent:main` key, shared-env two-session isolation, non-git cwd (null git meta), and DB-failure resilience.

## How to Test

1. Run a gateway with any messaging platform (`hermes gateway run`).
2. In a chat, ask the agent to `cd` into a git repo and run `pwd`.
3. Inspect the row: `sqlite3 ~/.hermes/state.db "SELECT cwd, git_repo_root, git_branch FROM sessions WHERE session_key LIKE 'agent:%' ORDER BY started_at DESC LIMIT 1"` → shows the repo, not the launch dir. The chat now appears under that repo in the Projects sidebar.

Automated: `scripts/run_tests.sh tests/tools/test_terminal_persist_session_cwd.py tests/state/test_update_session_cwd_by_session_key.py`

### Code

- [x] I've run `pytest tests/ -q` and all tests pass *(full suite via `scripts/run_tests.sh`: all cwd/state suites green; the ~50 remaining failures are pre-existing on clean `main` and environmental — Daytona/Modal/fal/ACP deps — each baseline-verified)*
- [x] I've tested on my platform: Linux 6.8 (Ubuntu)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A *(docstrings added; no user-facing docs describe the sidebar grouping's data source)*
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — or N/A *(no new config; no new `HERMES_*` env vars)*
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — git is invoked via `subprocess` list args (no shell, no POSIX-only constructs); mirrors the TUI path's probe
- [x] I've updated tool descriptions/schemas if I changed tool behavior — or N/A *(no tool schema/prompt changes; prompt caching untouched)*


andrexibiza (commented):
Reviewed exact head `e0bf0257b2edaa03ea6390ce4aa8e5f321d6313d` against exact current `main@4209d371aa1bb8840ce8447555bdd863a1a96c38` (the PR's recorded base is current main, so there is no base-drift ambiguity here). I read the state helper, the terminal call site, the new state + E2E tests, the existing session-reset/run-generation guards, the existing TUI cwd/git-metadata generation path, #29531 and its cwd/workspace lineage, and the live adjacent storage work. Exact-head CI is clean: CI `33267535725`, Nix `33267535219`, Docker `33267535231` all completed `success`.

There is one session-ownership blocker before I think this should land, plus one latency issue worth fixing while the path is still narrow.

### P1 — the durable write is keyed by the reusable route, not the session instance that produced the terminal effect

`update_session_cwd_by_session_key()` deliberately resolves `session_key` to **whichever row is open now**. That is correct for a caller whose effect is known to belong to the current owner, but `terminal_tool` performs this write *after* the command completes and passes only the stable routing key. A reset chain explicitly reuses that key across multiple durable session ids; the new test even models `row_old → row_open` under one key.

That creates the opposite race from the one documented in the helper. The docstring covers:

`resolve S1 → reset closes S1 / creates S2 → write S1`

and correctly notes that the old row receiving the late update is benign. It does **not** cover:

`terminal effect starts in S1 → reset closes S1 / creates S2 → stale S1 effect completes → resolve K → write S2`

In the second ordering the old conversation's cwd/git metadata is attached to the newly-created conversation before S2 has executed any terminal command. `git_metadata_generation` does not save this: the generation claim happens only *after* the helper has selected the wrong durable session row.

The repository already has the missing identity available at this boundary. `gateway.session_context` binds both `HERMES_SESSION_KEY` **and `HERMES_SESSION_ID`** into task-local ContextVars, explicitly documents that the context is copied into executor threads, and reset/run code treats the durable session id / run generation as the conversation boundary. The TUI sibling path follows the same stronger shape: it claims cwd on an exact `session_id`, receives that row's git-metadata generation, and only then allows enrichment for that exact object.

Required repair: carry/consume the exact durable `session_id` that commissioned this terminal effect (or an equivalent session-generation token) and call `update_session_cwd(session_id, ...)` against that owner. Do not rediscover ownership from `session_key` after the effect has completed. If the exact owner no longer exists/is no longer admissible, the stale completion should be a no-op for the successor rather than retargeting itself. I would add a regression that creates S1 under K, captures S1 as the terminal context, resets to S2 under the same K, then completes the S1 persistence path and proves **S2 remains untouched**. The current five resolver tests prove the current-open preference; they do not prove stale-completion fencing.

This is the same defect class the existing gateway generation guard is already preventing for `_running_agents`: a stable route identifies a lane, not the generation that owns a late side effect.

### P2 — the git probe is still on the terminal tool's return path

The PR description says the synchronous probe is “off the hot path” because the command already returned. The child command has returned, but `terminal_tool()` has not: `_probe_git_meta(new_cwd)` runs inside `persist_gateway_session_cwd_to_db()` before output extraction/return to the agent. It can run two sequential `git rev-parse` subprocesses with 2-second timeouts, and this happens after every qualifying terminal command even when the cwd did not change.

That means a slow/wedged repository can add roughly four seconds to an otherwise completed tool call. This does not corrupt state, but it is avoidable and materially changes terminal latency. The existing TUI path already separates the authoritative cwd claim from git enrichment; reusing that pattern (or at minimum probing only when the authoritative cwd actually changes) would keep the Projects projection from becoming tool-call latency.

### Interlocks / lineage

I do **not** read this as a duplicate of the older cwd work. #19292 (SHL0MS) is the earlier open per-session cwd-storage proposal; closed #21427 (gogadoro0524) explored resume consumption; closed #17839 (zxd-666) addressed the different API/request-workspace transport half. #29531 is broader than this PR and its discussion explicitly separates per-request workspace authority from persisted session metadata, so “gateway part” is the right closure language rather than claiming the whole issue.

#97309 is complementary, not superseded: it is about selecting the profile-owned `state.db` instead of trusting ambient scope. This PR currently opens `Path(get_hermes_home()) / "state.db"` directly. On the normal inbound turn path that scope is expected to be installed, so I am not calling that a second blocker on this head; however, the exact-session-id repair above should compose with #97309's owner-store rule rather than introducing a new key/ambient bypass if that storage routing lands first.

What is already good here: the transient `workdir` and `cwd_observed` guards are preserved, DB failure is correctly non-fatal to the user's command, `update_session_cwd()` already clears stale repo identity when cwd actually moves, the test matrix covers ordinary reset-chain selection / non-git / failure behavior well, and this exact object is fully green. The missing piece is narrow but important: **the effect must stay attached to the session generation that produced it.** Once that is fenced, this is a solid fix for the Projects-sidebar persistence gap.

narigondelsiglo:
Both review points addressed in bd79e11605 (rebased on current main):

**P1 — session-generation fencing.** The key-based resolver (`update_session_cwd_by_session_key`) is removed entirely. Ownership is now carried by the task-local durable `HERMES_SESSION_ID` (bound by `gateway.session_context`, copied into executor threads) and written through `update_session_cwd(session_id, ...)`, which no-ops when the owner row is gone. A stale effect from a reset conversation therefore lands on its own (closed) row or is dropped — it can never re-home its successor. Regression added: S1 completes a cwd persistence after the reset minted S2 under the same key → S2 stays untouched, S1 keeps its own cwd.

**P2 — probe off the return path.** `record_session_cwd()` now returns the previous record, so the git probe (and the DB write) only run when the authoritative cwd actually changed. The first command of a session always persists; subsequent commands in the same directory skip both the probe and the write entirely.

Also dropped the resolver's test file (its five tests asserted the current-open-preference model that was the defect); the fencing test now covers the property the review asked for. Local run: 10/10 in the PR's test file + 40/40 in the neighboring cwd suites (cwd store, interrupted-command cwd, terminal task cwd, TUI cwd follow).

CI: https://github.com/NousResearch/hermes-agent/actions/runs (push event on the fork branch).

andrexibiza (commented):
Reviewed exact head `d5ac6eb4bbc90a48db9b659a15c63e73a22ad11a` against current `main@b20cc5f787ea816ea8645603b7b2ac8234dcb8b4` and the exact-head CI run.

The prior database-owner bug is fixed at the final write boundary: `persist_gateway_session_cwd_to_db()` now targets the task-local durable `HERMES_SESSION_ID` rather than re-resolving the currently open row from a reusable routing key. The prior performance finding is only partially fixed. This head is still not merge-ready.

## Blocking findings

### P1 — the exact-session fence is applied after reusable route state has already been mutated

The real completion path still runs:

```python
prev_cwd = record_session_cwd(session_key, observed_cwd)
...
persist_gateway_session_cwd_to_db(session_key, observed_cwd, ...)
```

`record_session_cwd()` writes `_session_cwd[session_key]`, while `/new` reuses that same routing key across durable session IDs. A late S1 completion can therefore overwrite the route-scoped cwd after S2 has been minted. The database write correctly remains owned by S1, but S2's next terminal command resolves its execution cwd from the stale `_session_cwd[K]` record and can run in S1's directory. The exact-ID fence is too late to protect runtime state.

The new stale-completion test does not exercise this path: after minting S2 it calls `persist_gateway_session_cwd_to_db()` directly, bypassing `record_session_cwd()` and the mutation that creates the bug.

The same ownership mismatch causes a non-overlap failure after reset. If S2's first command finishes in the same cwd S1 last recorded, `prev_cwd == observed_cwd`, so `probe_git=False`. S2's fresh row is then initialized with null git metadata, and every later command in that cwd continues to skip the probe.

Required closure: generation-bind the runtime/dedupe record to the exact durable session ID, or otherwise reject stale generations before mutating the reusable route record. Add full `terminal_tool()` regressions for (1) S1 completing after S2 owns the route and (2) S2 starting in the same cwd as S1 while still receiving branch/root metadata.

### P1 — git metadata is probed on the gateway host, not in the active terminal backend

`observed_cwd` is reported by the selected environment, but `_probe_git_meta()` invokes host-local `subprocess.run(["git", "-C", cwd, ...])`. That is only valid for the local backend. For Docker, SSH, Modal, Daytona, Singularity, and Vercel Sandbox, the cwd belongs to the container or remote host.

The common result is false null metadata, leaving project grouping to fall back to the raw cwd and preserving the original subdirectory-fragmentation bug. Worse, if the gateway host happens to contain the same path spelling, the PR can persist branch/root metadata for an unrelated host repository.

Required closure: execute the probe in the same environment that produced `observed_cwd`—through the environment abstraction or an in-band result marker—and add a backend test where the repository exists only in the execution environment plus a host-path-collision test. The current suite uses only `LocalEnvironment`, so it cannot validate the advertised messaging-gateway behavior across supported terminal backends.

### P1 — the new synchronous probe bypasses the repository's Windows-safe git subprocess contract

`tui_gateway.git_probe.run_git()` deliberately uses `hermes_cli._subprocess_compat.bounded_git_probe`; its contract documents the existing Windows failure where plain `subprocess.run(timeout=..., capture_output=True)` can remain blocked after killing git because a descendant retains the pipe handles. `_probe_git_meta()` reintroduces exactly that primitive, twice, on the synchronous terminal response path.

The PR description says this mirrors the TUI probe and is “off the hot path.” It does neither: the TUI uses the bounded helper, and `terminal_tool()` has not returned to the user until both probes and the database write finish.

Required closure: use the canonical bounded git process control for local execution and an equivalently bounded backend-native path for remote/container execution. Add the relevant timeout/cleanup regression rather than relying on `subprocess.run(timeout=2)`.

### P2 — unchanged-cwd commands still open and write `state.db` every time

The call to `persist_gateway_session_cwd_to_db()` is unconditional for every qualifying command; only `probe_git` is gated by `cwd_moved`. That contradicts both the call-site comment and the follow-up claim that unchanged-cwd commands skip “the git probe (and DB write).”

This is not a harmless no-op. Every call constructs a new `SessionDB`, performs a write, and `SessionDB.update_session_cwd()` increments `git_metadata_generation` even when cwd is unchanged. That creates per-command WAL/write amplification and rejects any legitimate in-flight metadata publisher carrying the previous generation despite no workspace move.

Required closure: suppress the write when the exact durable session already owns the same persisted state, while still allowing a newly minted durable session under the same routing key to perform its first write. Add a regression asserting a second same-cwd command does not issue another update and does not advance `git_metadata_generation`.

### P2 — linked worktrees are persisted as separate projects

`_probe_git_meta()` stores `git rev-parse --show-toplevel` directly as `git_repo_root`. That is the linked worktree's own root. The existing Projects contract intentionally folds linked worktrees under their common repository root using `--git-common-dir` (`tui_gateway.git_probe.common_repo_root()` / `_git_common_repo_root_for_cwd()`), and `workspace_key()` prefers the stored `git_repo_root` over cwd.

As written, a gateway session in each linked worktree is grouped as a separate project. Detached HEAD also persists the literal `HEAD` rather than following the canonical branch-or-short-commit fallback.

Required closure: reuse the canonical common-root and branch semantics in the backend-aware probe, with a real linked-worktree regression.

## Exact-head verification

CI run: https://github.com/NousResearch/hermes-agent/actions/runs/33431771721

- The new `tests/tools/test_terminal_persist_session_cwd.py` passes all 10 tests on this head.
- `Check contributors / check-attribution` is red because `narigondelsiglo@users.noreply.github.com` is unmapped: https://github.com/NousResearch/hermes-agent/actions/runs/33431771721/job/99618647395
- The Python test job is red on `tests/tools/test_browser_lightpanda.py::TestChromeFallback::test_chrome_fallback_injects_required_sandbox_args`; this appears unrelated to the two changed files, but the exact head remains red: https://github.com/NousResearch/hermes-agent/actions/runs/33431771721/job/99618648171
- The OSV job reached SARIF upload and failed on GitHub's installation API rate limit: https://github.com/NousResearch/hermes-agent/actions/runs/33431771721/job/99618575212
- Nix flake check and Docker Build, Test, and Publish are green.

At review time the branch is also 54 commits behind current `main`. Rebase before final verification, resolve the deterministic attribution failure, rerun transient infrastructure failures, and require the complete exact-head check set to be green.

## PR description drift

The body still describes a `hermes_state.py` helper and `tests/state/test_update_session_cwd_by_session_key.py`, neither of which exists in the current diff. It also says there are 9 E2E tests, provides an automated command naming the removed state test, and claims unchanged-cwd commands skip the DB write. Update the description and verification claims to match the final implementation after the blockers above are closed.

andrexibiza (commented):
Re-reviewed exact head `36c7dceaab33148646f42aa60a8bc58d7efa9fd3` against current `main@21b2095d00a98b8ad7b5c60b10587619c852cdb8`. The current head itself is green in CI `33468862737`, Docker `33468862346`, and Nix `33468862352`, and GitHub reports the PR mergeable. The second-round blockers from the prior review are still present in the live diff, so this is not ready to land yet.

### P1 — stale session generations can still mutate the reusable runtime cwd before the exact-ID fence

The durable DB write is now keyed to `HERMES_SESSION_ID`, but the real completion path still calls `record_session_cwd(session_key, observed_cwd)` first. `_session_cwd` is keyed only by the reusable routing key. A late S1 completion after `/new` has minted S2 can therefore overwrite `_session_cwd[K]` with S1's cwd even though the later DB write correctly stays on S1. S2's next terminal command can then inherit the stale S1 working directory.

The regression still bypasses the offending mutation: after minting S2 it calls `persist_gateway_session_cwd_to_db()` directly rather than driving the full `terminal_tool()` completion path. The runtime record itself needs generation ownership (exact durable session ID or equivalent) or a stale-generation check *before* `record_session_cwd()` mutates the route-scoped cache. Add a full-path stale-completion witness and the same-cwd-new-session witness; the latter must prove S2 still receives its own git metadata even when S1 previously recorded the same cwd under K.

### P1 — git metadata is still probed on the gateway host instead of the active terminal backend

`_probe_git_meta(cwd)` still runs host-local `subprocess.run(["git", "-C", cwd, ...])`. `observed_cwd` belongs to the selected `Environment`. For SSH, Docker, Modal, Daytona, Singularity, Vercel Sandbox, etc., that path may not exist on the gateway host—or worse, may name an unrelated host checkout. That can persist null or wrong `git_repo_root`/branch for a valid remote session.

The probe must execute in the same environment that produced the cwd (or consume metadata emitted in-band by that environment), with tests for a repo that exists only in the backend and a host-path collision.

### P1 — the local probe still bypasses the repository's bounded git-process owner

The diff still introduces two plain `subprocess.run(..., timeout=2, capture_output=True)` git probes. Current main already centralizes this class in `hermes_cli._subprocess_compat.bounded_git_probe`; `tui_gateway.git_probe` explicitly documents why plain `subprocess.run(timeout=...)` can deadlock on Windows when descendants retain pipes. Reuse the bounded owner for local execution and provide equivalently bounded process settlement for backend-native execution.

### P2 — unchanged-cwd commands still write `state.db`

The call-site comment says later same-cwd commands skip both probe and DB write, but the code calls `persist_gateway_session_cwd_to_db(...)` unconditionally; `probe_git=False` suppresses only the git subprocesses. That still constructs a `SessionDB`, calls `update_session_cwd()`, and advances metadata generation on every qualifying command. Gate the durable write itself on whether this exact durable session already owns the persisted cwd, while preserving the first write for a newly minted S2 even if its route key's cached cwd equals S1's.

### P2 — project identity still diverges from the canonical Projects semantics

`git rev-parse --show-toplevel` stores a linked worktree root as a distinct project. Existing Projects logic uses the common repository root and canonical detached-HEAD fallback. The backend-aware probe should reuse those semantics so gateway sessions do not fragment linked worktrees into separate Projects or persist literal `HEAD` as the branch.

### Release-object gates

The current submitted train contains two commits: `9c43461f76fcd0184265b682a3c2acc7166bb4dc → 36c7dceaab33148646f42aa60a8bc58d7efa9fd3`. The head has the green matrix above, but the first surviving commit has no PR-triggered workflow runs, so the every-surviving-commit gate is not satisfied. Also, `tools/terminal_tool.py` is still above the repository's 2,000-line ceiling (line 2001 exists on this exact head); this change adds a new persistence/git-probe responsibility directly inside that godfile instead of extracting a bounded owner.

The PR body is also stale relative to the current two-file diff: it still describes the removed `hermes_state.py` helper/state test and claims behavior that the current call site does not implement. Update it only after the product shape is final.

No new issue class is needed here; these are the same unresolved ownership/backend/process-lifetime/write-amplification blockers already identified, now verified against the live exact head.

PR #80243, opened 2026-08-06
fix(recovery): sqlite_sequence hint caps the high-edge probe when the tail leaf is corrupt (#80205)

thatssoheil:
## Summary

Fixes #80205 — the practical high-edge bound still fails on main for the exact corruption shape the issue describes.

**Context:** main merged `_probe_populated_edge` (6dad74596), but its outward gallop cannot confirm emptiness when the damaged leaf sits at the edge of the populated range: every `rowid > ?` probe must cross the corrupt page and raises `database disk image is malformed`, so the gallop doubles into the synthetic tail and `high` stays at the 64-bit domain max — the same budget-exhaustion, tail-row-skipped failure #80205 reported. Live-reproduced on current main.

**Fix (surgical, on top of main's implementation):** `_probe_populated_edge` now consults `sqlite_sequence` first — the issue's suggestion 2 — for the `high` edge. AUTOINCREMENT tables (the normal shape for session/message tables) record the highest assigned rowid there, so the upper bound is capped in one query instead of 63 doomed probes. The hint is strictly best-effort: only honored when `seq >= anchor`, skipped on any read error, and never the sole source of truth.

## Verification

- New regression `test_allow_partial_reaches_tail_row_when_high_edge_probe_fails`: corrupts the rightmost data leaf (descending edge probe raises), asserts the practical bound lands near the 320-row populated range and no rows are skipped.
- RED on pre-fix main: `high=9223372036854775807` for 320 rows.
- GREEN with the hint; full `tests/hermes_cli/test_session_recovery.py` passes (5/5).

Issue #80205, opened 2026-08-06
[Bug]: sessions recover --allow-partial can exhaust its range-query budget and omit a recoverable tail row


mzkarami:
## Bug Description

A preserved, damaged `state.db` exposed a gap in the offline recovery path on Hermes Agent v0.20.0.

`hermes sessions recover --inspect-only` correctly reported that the canonical `sessions` and `messages` tables were not *completely* readable. The documented `--allow-partial` path then produced a verified output database, but its rowid-range salvage exhausted the 10,000-query budget after the high-rowid probe failed. It recovered 76,881 messages even though SQLite's page-level `.recover --ignore-freelist` recovered 76,882 contiguous message rows from the same stopped source bundle.

The recovery report shows why:

```text
messages.rowid_bounds.low: 1
messages.rowid_bounds.high: 9223372036854775807
messages.rowid_bounds.errors: ["high rowid: database disk image is malformed"]
messages.rowid_bounds.fallback_edges: ["high"]
messages.query_limit_reached: true
messages.copied_rows: 76881
messages.skipped_rowid_ranges:
  - low: 76882
    high: 81848
    error: database disk image is malformed
  - low: 81849
    high: 9223372036854775807
    error: salvage range query limit reached
```

The actual recoverable message ID range was `1..76882`. The domain-wide fallback therefore spent the query budget bisecting an enormous range that could not contain real rows and omitted recoverable message ID 76882.

This is separate from the runtime corruption reports: this issue is specifically about maximizing safe, offline recovery after a source has already been preserved and all writers have been stopped.

## Steps to Reproduce

This was reproduced against a preserved corrupt source bundle; the active database was never opened or modified by the recovery command.

1. Use Hermes Agent v0.20.0 with a damaged `state.db` where:
   - the low rowid edge is readable;
   - `SELECT rowid FROM messages ORDER BY rowid DESC LIMIT 1` raises `database disk image is malformed`;
   - the final canonical message row is still recoverable at the page level.
2. Confirm strict inspection fails:

```shell
hermes sessions recover \
  --source preserved-state.db \
  --inspect-only
```

Relevant result:

```text
recoverable: false
required table sessions is not completely readable
required table messages is not completely readable
```

3. Run the documented partial salvage path:

```shell
hermes sessions recover \
  --source preserved-state.db \
  --output partial-recovered.db \
  --allow-partial
```

[...]

PR #98062, opened 2026-08-29
fix(recovery): try min/max rowid aggregates before the domain fallback

### LINKED ISSUE #98050: sessions recover --allow-partial: damaged low edge of a 4-row table exhausts the salvage budget over the full 64-bit rowid domain, though min(rowid)/max(rowid) answer the bounds in one query
## Bug Description

Production corruption on hermes-agent 0.20.6 exposed two gaps in the offline recovery path. This report is based on a real damaged `state.db` bundle (~899 MiB main + ~124 MiB WAL); all numbers below were cross-checked manually with `sqlite3` on preserved copies of the bundle. No user content is included.

**Gap 1 (primary): rowid-range salvage exhausts its 10,000-query budget bisecting the full 64-bit domain on a tiny table when the *low* edge is damaged — even though `min(rowid)`/`max(rowid)` aggregates return the true bounds in one query.**

`gateway_routing` (4 rows) sustained B-tree damage. Its partial-recovery report:

```json
"gateway_routing": {
  "copied_rows": 0,
  "error": "2 rowid range(s) skipped",
  "mode": "rowid_range_salvage",
  "query_limit_reached": true,
  "range_queries": 10000,
  "rowid_bounds": {
    "errors": ["low rowid: database disk image is malformed"],
    "fallback_edges": ["low"],
    "high": 4,
    "low": -9223372036854775808
  },
  "skipped_rowid_ranges": [
    { "error": "database disk image is malformed",
      "low": -9223372036854775808, "high": -9223372036854772494 },
    { "error": "salvage range query limit reached",
      "low": -9223372036854772493, "high": 4 }
  ],
  "source_rows": 4,
  "status": "failed"
}
```

The low-edge probe failed, the fallback anchor went to `INT64_MIN`, and the bisection burned the entire budget while moving the low bound inward by only ~3.3e3 rowids out of a 9.2e18-wide domain. All 4 rows were skipped and the table was lost from the output.

Manual verification on the same preserved bundle (sqlite3 3.45.1):

```sql
SELECT count(*) FROM gateway_routing;              -- 4  (succeeds)
SELECT min(rowid), max(rowid) FROM gateway_routing; -- 1 | 4  (succeeds, one query)
SELECT scope, session_key FROM gateway_routing;     -- all 4 rows (succeeds)
SELECT length(entry_json) FROM gateway_routing;     -- fails: database disk image is malformed
```

So the true rowid bounds that salvage spent 10,000 queries failing to establish were available from a single aggregate. The rows' short columns were readable; only the long `entry_json` payload (which crosses the damaged pages) was not.

This is the low-edge mirror of #80205, but the proposed fix there cannot cover it: PR #80243 caps the **high** edge via `sqlite_sequence`, while `gateway_routing` has no `AUTOINCREMENT` (`PRIMARY KEY (scope, session_key)`), and the missing bound here is the **low** one. There is also no equivalent of the "populated range" heuristic near `INT64_MIN` — every probe between `INT64_MIN` and the first real rowid is wasted work by construction for any Hermes table, whose rowids start at 1.

**Gap 2 (secondary): `--inspect-only` does not validate TEXT payload decodability, so it reported `recoverable: true` with `errors: []` / `warnings: []` on a bundle containing an undecodable row.**

`--inspect-only` output (abridged): all canonical tables `available`, `messages: 79895` rows, `recoverable: true`, no errors, no warnings. Strict recovery then aborted mid-copy at message rowid 79,001 with:

```
Could not decode to UTF-8 column 'tool_calls' with text '[{"id": "call_…'
```

i.e. one row's `tool_calls` TEXT payload contained invalid UTF-8 (the same row's REAL `timestamp` had also been physically corrupted to a denormal `2.5e-308`). `--allow-partial` handled it correctly — it skipped exactly that row (79,894/79,895 recovered, `query_limit_reached: false`) and continued to a verified output. But the inspect report's clean bill of health did not predict strict mode aborting 895 rows early, which reads to a user like "895 rows are lost" when the actual loss is 1 row.

This is distinct from #86224: that issue is about a covering-index `COUNT(*)` masking structurally damaged payload *pages* (strict then fails wholesale). Here the pages traverse fine and counts are accurate; what inspection misses is per-row column-level decodability, which is what strict/partial recovery actually operate on.

## Environment

- hermes-agent 0.20.6, gateway running against `~/Hermes/home/state.db` (WAL mode)
- Ubuntu 24.04, sqlite3 3.45.1; DB ~899 MiB main / ~124 MiB WAL at time of failure
- `PRAGMA integrity_check` on the preserved bundle reported damage across the `messages` table, `gateway_routing` table, four `sessions` secondary indexes, and `messages_fts_data` (101 findings, e.g. `btreeInitPage() returns error code 11`, `invalid page number`, `2nd reference to page`), consistent with the "canonical B-tree damage misreported as FTS corruption" class already tracked in #88587 / #97940

## Outcome for context

With the two gaps worked around manually (4 routing entries were reconstructible from the sidecar `sessions/sessions.json`), the final output database passed `integrity_check`, FK checks, and both FTS rebuilds; all 4,172 sessions and 79,894/79,895 messages were restored, sacrificing only the physically-damaged row. The recovery JSON report quality (per-table status, rowid bounds, skipped ranges) is what made the manual repair possible — genuinely good instrumentation.

## Suggestions

1. **Seed the rowid bounds with aggregates before bisection**: probe `SELECT min(rowid), max(rowid) FROM t` first; honor the result when it succeeds (it did here, in one query), fall back to edge probes only on error. A conservative fallback anchor of `1` (instead of `INT64_MIN`) for the low edge would also have bounded this case, since Hermes canonical tables assign positive rowids.
2. **Fail over to a plain scan for small tables**: if `count(*)` (or a bounded direct scan) succeeds and row count is small, copy directly instead of entering rowid-range salvage; salvage should only engage when a direct scan fails.
3. **Make `--inspect-only` validate what recovery reads**: include a bounded payload-decodability probe (e.g. attempt the same projection strict mode uses, chunked) so "recoverable: true" distinguishes "pages traverse" from "every row will decode"; and have the strict-mode report state explicitly that copying stopped at the first undecodable row, with the count of rows *after* the stop that partial mode went on to recover.
4. Optionally: when a table's salvage fails, consult `sessions/sessions.json` as a reconstruction source for `gateway_routing` before giving up (it held all 4 entries in full in our case).

Happy to provide additional report excerpts or run proposed patches against the preserved bundle — the original is intact and hash-pinned.

liuhao1024:
## What does this PR do?

Fixes Gap 1 of #98050: when `sessions recover --allow-partial` runs rowid-range salvage on a table whose table b-tree edge is damaged, the ordered rowid-edge probes fail and the missing edge falls back to the full 64-bit rowid domain. Range bisection then burns the entire 10,000-query budget subdividing a synthetic tail that cannot contain rows, and readable rows are recorded as skipped (the report's `gateway_routing` case: 4 real rows, `low = INT64_MIN`, `query_limit_reached: true`, table lost).

The fix tries `SELECT min(rowid), max(rowid)` once after an ordered probe fails. The ordered probes walk the table b-tree, but the aggregate can satisfy itself from any covering index — an independent physical path through the file, which is exactly why a damaged edge can kill one and not the other. When the aggregate answers, it returns the **true** bounds in one query, so:

- clean databases are untouched (the aggregate only runs after a probe fails, and only non-NULL values are adopted);
- edges recovered this way are real bounds, not guesses, so they stay out of `fallback_edges` and skip the gallop probing entirely;
- if the aggregate fails too (e.g. no covering index), behavior is unchanged (gallop / domain fallback).

## Related Issue

Fixes #98050 (Gap 1 — the rowid-boundary half; Gap 2, `--inspect-only` payload decodability, is not addressed here)

## Type of Change

## Changes Made

- `hermes_cli/session_recovery.py` — `_salvage_rowid_bounds`: after an ordered edge probe fails, try `SELECT min(rowid), max(rowid)` before the unavailable/domain-fallback paths; record recovered edges in `rowid_bounds.aggregate_recovered` and append an error entry if the aggregate itself raises.
- `tests/hermes_cli/test_session_recovery.py` — two regression tests: (1) physical root-header damage on a table with a covering index → both ordered probes raise while the aggregate answers `[1, 40]`, and the bounds use those true values with no `fallback_edges`; (2) no covering index → the aggregate fails too and the table stays `unavailable` instead of inventing bounds.
- `tests/hermes_cli/test_session_recovery_lost_and_found.py` — the #80205 high-edge test now asserts the aggregate path (the damaged high edge is answered by the covering-index aggregate, so `fallback_edges` is empty); all downstream row-recovery assertions are unchanged and still pass.

## How to Test

1. `pytest tests/hermes_cli/test_session_recovery.py tests/hermes_cli/test_session_recovery_lost_and_found.py -q` — Observed result: `13 passed` (on unmodified `upstream/main`: `11 passed`; the new file adds 2 tests, and 1 existing premise assertion moves from the gallop path to the aggregate path with the same recovery outcome).
2. `pytest tests/test_hermes_state.py -q` — Observed result: `243 passed, 2 skipped`.
3. Physical premise check (embedded in test 1): with the table b-tree root damaged and a `PRIMARY KEY (scope, session_key)` covering index intact, `SELECT rowid ... ORDER BY rowid ASC/DESC LIMIT 1` raises `database disk image is malformed` while `SELECT min(rowid), max(rowid)` returns the true `(1, 40)` — the covering-index plan is a genuinely independent path, matching the manual sqlite3 verification in the issue.

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate (#80243 caps the high edge via the `sqlite_sequence` hint for AUTOINCREMENT tables; this PR covers tables without a sequence entry, which that PR's own notes call out as the residual gap)
- [x] I've run `pytest tests/ -q` and all tests pass (full targeted suites: session_recovery ×13, hermes_state ×243)
- [x] I've tested on my platform: macOS 26.6 (arm64)

- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) (pure-stdlib sqlite3 logic, no platform-specific code)

## For New Skills

N/A — not a skill.

## Screenshots / Logs

Not applicable — behavior is observable via the recovery JSON report fields (`rowid_bounds.aggregate_recovered`, empty `fallback_edges`) asserted in the tests.

### DIFF
--- a/hermes_cli/session_recovery.py
+++ b/hermes_cli/session_recovery.py
@@ -493,6 +493,29 @@ def _salvage_rowid_bounds(
     if rows["low"] is None and rows["high"] is None and not result["errors"]:
         result["empty"] = True
         return result
+
+    # Issue #98050: an ordered edge probe walks the table b-tree, but the
+    # min/max aggregates can satisfy themselves from any covering index —
+    # an independent physical path through the file. When damage eats the
+    # table b-tree edge the aggregate answers with the true bounds in one
+    # query, instead of leaving the missing side on the 64-bit domain
+    # fallback that range bisection spends its whole budget subdividing.
+    if rows["low"] is None or rows["high"] is None:
+        try:
+            agg = source.execute(
+                f'SELECT min(rowid), max(rowid) FROM "{table}"'
+            ).fetchone()
+        except sqlite3.DatabaseError as exc:
+            result["errors"].append(f"min/max rowid aggregate: {exc}")
+            agg = None
+        if agg is not None:
+            if rows["low"] is None and agg[0] is not None:
+                rows["low"] = int(agg[0])
+                result.setdefault("aggregate_recovered", []).append("low")
+            if rows["high"] is None and agg[1] is not None:
+                rows["high"] = int(agg[1])
+                result.setdefault("aggregate_recovered", []).append("high")
+
     if rows["low"] is None and rows["high"] is None:
         result["unavailable"] = True
         return result
--- a/tests/hermes_cli/test_session_recovery.py
+++ b/tests/hermes_cli/test_session_recovery.py
@@ -17,6 +17,7 @@ from hermes_cli import session_recovery
 from hermes_cli.session_recovery import (
     SessionRecoverySafetyError,
     SessionRecoverySourceError,
+    _salvage_rowid_bounds,
     inspect_session_database,
     recover_session_database,
 )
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** Fixes rowid-range salvage burning its 10k query budget on a synthetic 64-bit tail when the table b-tree edge is damaged.

**What it does**
- `hermes_cli/session_recovery.py:16` — when `rows["low"]` or `rows["high"]` is still `None` after ordered edge probes, tries `SELECT min(rowid), max(rowid) FROM "table"` (one query). This can be satisfied from a covering index on a different physical page, while ordered `ORDER BY rowid LIMIT 1` walks the damaged table b-tree (`hermes_cli/session_recovery.py:18`).
- Recovered edges are stashed in `rows["low"|"high"]` and `result["aggregate_recovered"]`; failure appends to `result["errors"]` and falls through to `unavailable`.
- Tests `tests/hermes_cli/test_session_recovery.py:66` / `:132` flip a b-tree page header (`_damage_btree_root_header`) to verify both probe-fails-but-aggregate-succeeds and aggregate-also-fails → `unavailable`.

**Non-blocking notes**
- Quote of table name as `"{table}"` is safe here — `table` comes from `sqlite_master` name enumeration, not user input.
- No new query budget cost: one aggregate replaces up to 10k bisection probes on the synthetic tail.

No functional issues found.

liuhao1024:
Thanks for the review. Right on both notes — the quoted `"{table}"` identifier comes from `sqlite_master` enumeration rather than user input, and the single `min/max` aggregate replaces up to 10k bisection probes on the synthetic tail, so the query budget is strictly reduced.

ehz0ah (commented):
Reviewed exact head `a5098e74270e35dfe4317668fa92311d1978f995`.

The change is correct for the reported case. When ordered rowid edge probes fail on a damaged table B-tree, a covering index can still answer `min(rowid), max(rowid)`. The recovered values are treated as real bounds, so salvage avoids the synthetic 64-bit tail and does not spend its query budget before reaching readable rows. If the aggregate also fails, the table remains unavailable instead of inventing bounds.

I ran the two targeted recovery files through `scripts/run_tests.sh`: 13 tests passed. Ruff also passed, and the PR's CI is green. I found no correctness blocker at this head.

This PR is now superseded by #108130, which landed the same aggregate-bound recovery on current `main` as commit `04dd80a977f40b05e5b2054111747af07a61886a`, together with later recovery changes.

PR #95162, opened 2026-08-26
fix(doctor): drop --json authenticated for gh CLI 2.98+ compatibility

chuenlye:
## Problem

`gh auth status --json authenticated` fails on gh CLI 2.98+ because the `authenticated` JSON field was removed (only `hosts` remains). The command exits with code 1 and prints:

```
Unknown JSON field: "authenticated"
Available fields:
  hosts
```

This causes `_gh_authenticated()` in `hermes_cli/doctor.py` to return `False` even when the user is logged in via `gh auth login`, so the doctor falsely reports:

> ⚠ No GITHUB_TOKEN (60 req/hr rate limit)

## Fix

Drop `--json authenticated` and use plain `gh auth status`. The code only checks `returncode == 0` and never parses stdout, so `--json` was unnecessary. Plain `gh auth status` works across all gh versions and returns exit 0 when authenticated.

## Verification

- Tested with gh 2.98.0 (2026-08-20): `gh auth status` → exit 0 ✓, `gh auth status --json authenticated` → exit 1 ✗
- Doctor now correctly shows: `✓ GitHub authenticated via gh CLI`
- Existing tests only assert `cmd[:2] == ['gh', 'auth']`, so they remain green

PR #16132, opened 2026-04-26
fix(doctor): check gh auth status when GITHUB_TOKEN absent

jjjojoj:
## What
hermes doctor showed 'No GITHUB_TOKEN (60 req/hr rate limit)' warning even when the user had authenticated via gh auth login (stores token in ~/.config/gh/hosts.yml).

Fix: added _gh_authenticated() helper that runs gh auth status --json authenticated. When GITHUB_TOKEN and GH_TOKEN are both absent, falls back to this check — shows 'GitHub authenticated via gh CLI (full API access — no GITHUB_TOKEN needed)' if authenticated.

## Test
Added TestGitHubTokenCheck with 3 cases:
- no token + no gh binary → shows warning
- GITHUB_TOKEN env var present → shows ok
- no token + gh authenticated → shows ok via gh fallback

Fixes #16115

PR #98063, opened 2026-08-29
fix(doctor): gh CLI auth is detected on every gh version (#98051, salvage #95162)

### LINKED ISSUE #98051: doctor still warns 'No GITHUB_TOKEN' when gh < 2.52 (--json unsupported) — fix from #16115 regresses on older gh CLI
## Bug Description

`hermes doctor` still shows `⚠ No GITHUB_TOKEN (60 req/hr rate limit …)` even though the user is authenticated via `gh auth login`, on systems where the installed `gh` CLI predates v2.52.0.

The fix for #16115 (PR #16132, merged as `103f51a`) added a `_gh_authenticated()` fallback in `hermes_cli/doctor.py`, but it relies on `gh auth status --json authenticated`. The `--json` flag on `gh auth status` was only introduced in **gh CLI v2.52.0** (Oct 2024). On older `gh` versions (e.g. Ubuntu 24.04 "noble" ships `gh` 2.45.0), `gh auth status --json authenticated` fails with `unknown flag: --json` and exit code 1, so the fallback returns `False` and the original warning persists — i.e. the original bug is *not* actually resolved on such systems.

## Steps to Reproduce

1. Have a `gh` version < 2.52.0 (e.g. `gh 2.45.0` from Ubuntu). Verify:
   `gh --version` → `gh version 2.45.0 (2026-03-17 Ubuntu 2.45.0-1ubuntu0.3+esm3)`
2. Authenticate: `gh auth login` (token valid, `gh auth status` shows `✓ Logged in …`)
3. Do **not** set `GITHUB_TOKEN` / `GH_TOKEN` in the environment or `~/.hermes/.env` (that is exactly the intended gh-CLI-only setup).
4. Run `hermes doctor`.
5. Observe the warning:
   `⚠ No GITHUB_TOKEN (60 req/hr rate limit — set in ~/.hermes/.env for better rates)`

Direct proof the fallback cannot work on this `gh`:
```
$ gh auth status --json authenticated
unknown flag: --json

Usage:  gh auth status [flags]
  -h, --hostname string   Check a specific hostname's auth status
  -t, --show-token        Display the auth token
$ echo $?
1
```

## Expected Behavior

`hermes doctor` should recognize gh-CLI authentication on *any* supported `gh` version and print the success status `✓ GitHub authenticated via gh CLI` (as it does on gh ≥ 2.52.0), rather than the warning.

## Actual Behavior

The warning is shown whenever `gh` < 2.52.0, because `_gh_authenticated()` calls `gh auth status --json authenticated`, which is unsupported there.

## Root Cause

`hermes_cli/doctor.py`, `_gh_authenticated()` (currently ~line 3195):

```python
def _gh_authenticated() -> bool:
    """Check if gh CLI is authenticated via token file or device flow."""
    try:
        result = subprocess.run(
            ["gh", "auth", "status", "--json", "authenticated"],
            capture_output=True, timeout=10,
        )
        return result.returncode == 0
    except (FileNotFoundError, subprocess.TimeoutExpired):
        return False
```

The `--json` flag of `gh auth status` was added in gh CLI v2.52.0; on earlier versions the command exits 1, so `returncode == 0` is never true even when the user is fully authenticated.

## Suggested Fix

Make the detection version-agnostic. Either:
- Use `gh auth token` (exit code 0 = authenticated; no `--json` dependency), or
- Call plain `gh auth status` (exit code 0 already means "authenticated"), optionally keeping `--json` as a first attempt and falling back to the non-JSON form when the call fails with an unknown-flag error.

## Related Issues

- #16115 — original report (closed as completed by PR #16132); this issue is the regression gap in that fix.
- #39540 — adjacent robustness issue in the same `_gh_authenticated()` helper (PermissionError handling), also still uses `--json`.

## Environment

- OS: Linux 6.17.0-1031-nvidia (Ubuntu-based, DGX "Spark" GB10)
- Hermes: v0.20.6 (2026.8.27), install method `git` (upstream `main` still contains the `--json` code path)
- `gh` CLI: 2.45.0 (Ubuntu 2.45.0-1ubuntu0.3+esm3)
- `gh auth status`: authenticated (account `skbotoc1-web`)
- `GITHUB_TOKEN` / `GH_TOKEN`: unset (by design — gh-CLI auth only)

teknium1:
## Summary
`hermes doctor` now recognizes gh-CLI authentication on every gh version — the check invokes plain `gh auth status` instead of `gh auth status --json authenticated`, which fails on gh < 2.52 (`--json` flag not yet added) AND on gh 2.98+ (`authenticated` JSON field removed). Root cause: the #16132 fallback only checked `returncode == 0` and never parsed output, so the `--json` flag was pure downside.

Salvages PR #95162 by @chuenlye (authorship preserved). Fixes #98051.

## Changes
- `hermes_cli/doctor.py`: `_gh_authenticated()` drops `--json authenticated`; exit-code semantics are stable across all gh versions.
- `tests/hermes_cli/test_doctor.py`: regression test pinning the command shape (fails if `--json`/`authenticated` is ever re-added).
- `contributors/emails/`: attribution mapping for @chuenlye.

## Validation
- Targeted tests: `tests/hermes_cli/test_doctor.py -k "gh_auth or github or permission"` → 3 passed.
- Live repro on main: `gh auth status --json authenticated` → exit 1 on both gh 2.45 (`unknown flag`) and gh 2.98+ (`Unknown JSON field`); plain `gh auth status` → exit 0 when logged in.
- Repo sweep: only one production call site uses this shape; `web_routers/git.py` and `web_git.py` already use plain `gh auth status`.

## Infographic

doctor gh auth compat


alt-glitch:

Supersedes #95162 while preserving its authorship; this maintainer salvage applies the cross-version plain `gh auth status` repair for #98051.

PR #98065, opened 2026-08-29
test(cli): complete launchd simulation on Windows

### LINKED ISSUE #98038: [Bug]: launchd PID-exclusion tests call os.getuid on Windows
## Bug description

The three `TestServicePidSweepExclusion` cases in `tests/hermes_cli/test_update_launchd_unloaded_gateway.py` fail when the updater test matrix is run on native Windows. The fixture simulates macOS by forcing `is_macos()` to true, but the current launchd lookup reaches `os.getuid()`, which Windows Python does not provide.

This is a test portability defect, not a product launchd failure.

## Reproduction

Environment: Windows 11 build 26200, Python 3.11.7, pytest 9.1.1. Tested on `main@3f36c87e1e`; the implicated files are unchanged through current `main@835a913ffd`.

```powershell
.\.venv\Scripts\python.exe -m pytest -q --tb=short `
  tests/hermes_cli/test_update_launchd_unloaded_gateway.py::TestServicePidSweepExclusion
```

Actual result: three failures at `hermes_cli/gateway.py::_probe_launchd_domain_for_label` with `AttributeError: module 'os' has no attribute 'getuid'`.

## Root cause and control

The class patches `is_macos`, `get_launchd_label`, and `_launchd_domain`, but the path under test now probes a domain per label and reads `os.getuid()` directly. Its simulated platform boundary is incomplete.

As a causal control only, adding this fixture patch makes all three cases pass on Windows:

```python
monkeypatch.setattr(gateway_mod.os, "getuid", lambda: 501, raising=False)
```

Nearby `test_update_launchd_fleet_restart.py` already declares `pytest.mark.macos_only` because its helpers use POSIX `os.getuid()`.

## Expected behavior / closure criteria

- The canonical updater test matrix does not fail on Windows because a simulated macOS fixture accidentally calls an unavailable host API.
- The three service-PID exclusion contracts continue to execute on a platform where their simulated launchd boundary is complete.
- macOS coverage continues to exercise the real launchd helper path.

The outcome may be achieved by completing the simulation or by applying the appropriate platform scope; the issue does not require a particular implementation.

## Related work (not duplicates)

- #74973 and merged #92963 fix booted-out launchd gateway behavior. They do not cover native-Windows execution of this test fixture.
- #97633 / #97634 did not introduce the failure; it reproduces on clean `main`.

I found no existing issue or active PR for these three Windows failures.

sylvesterkaczmarek:
## Summary

- complete the simulated macOS launchd fixture used by `TestServicePidSweepExclusion`
- provide the mocked `os.getuid()` value that the launchd domain probe expects
- keep the three PID-exclusion contracts running on Windows instead of skipping them

## Root cause

The tests already simulate macOS by stubbing `is_macos()`, the launchd label, and the launchd domain. The current `_get_service_pids()` path also reaches `os.getuid()`, but native Windows Python does not expose that POSIX API.

That left the simulated platform boundary incomplete: the tests failed before reaching the launchd PID-exclusion behaviour they are meant to verify.

The fixture now supplies the same UID (`501`) as its mocked `gui/501` launchd domain. `raising=False` allows the attribute to be installed on Windows while remaining harmless on macOS.

## Testing

Independent Windows validation on exact `head@0a086f55d7` confirmed:

- on current `main@4209d371aa`, all three `TestServicePidSweepExclusion` cases fail because native Windows does not provide `os.getuid()`
- on this PR head, the same three cases pass
- the complete modified test file passes: `11 passed`

The reviewer found no active overlapping implementation and confirmed that the patch remains within the closure criteria of #98038.

Closes #98038

### DIFF
--- a/tests/hermes_cli/test_update_launchd_unloaded_gateway.py
+++ b/tests/hermes_cli/test_update_launchd_unloaded_gateway.py
@@ -153,6 +153,8 @@ class TestServicePidSweepExclusion:
         monkeypatch.setattr(gateway_mod, "is_macos", lambda: True)
         monkeypatch.setattr(gateway_mod, "get_launchd_label", lambda: "ai.hermes.gateway")
         monkeypatch.setattr(gateway_mod, "_launchd_domain", lambda: "gui/501")
+        # Complete the simulated macOS boundary on hosts without POSIX getuid.
+        monkeypatch.setattr(gateway_mod.os, "getuid", lambda: 501, raising=False)
 
         def fake_run(argv, **kwargs):
             if argv[:2] == ["launchctl", "list"]:


cervantesh:
Thanks for taking this.

For additional Windows validation, I ran the exact affected selection on current `main@4209d371aa` and on this PR’s exact `head@0a086f55d7`:

- `main`: the three `TestServicePidSweepExclusion` cases fail at `hermes_cli/gateway.py::_locate_launchd_gateway_service` because native Windows does not provide `os.getuid()`.
- PR head: the same three cases pass.
- PR head: the complete modified test file passes (`11 passed`).

The patch completes the fixture’s simulated macOS boundary while retaining the PID-exclusion tests on Windows rather than skipping them. I found no active overlapping implementation, and the change remains within the strict closure criteria of #98038.

Based on this evidence, `Closes #98038` is supported.

monerostar:

Host: Windows 11 (build 26200.9168), Python 3.11.15, pytest 9.1.1. `os.getuid` is absent on this interpreter (`AttributeError: module 'os' has no attribute 'getuid'`).

Reproduced on a current tree (`ha-wt-99245`, same fixture as this PR minus the two lines), then applied the patch locally.

**Before** (unpatched fixture):

```
3 failed, 8 passed
TestServicePidSweepExclusion.test_list_failure_falls_back_to_domain_print
TestServicePidSweepExclusion.test_both_interfaces_negative_means_no_pid
TestServicePidSweepExclusion.test_registered_but_not_running_has_no_pid_line
```

Failure site is exactly the one this PR names:

```
hermes_cli/gateway.py:1920 in _locate_launchd_gateway_service
    uid = os.getuid()  # windows-footgun: ok — POSIX launchd (macOS) helper, never invoked on Windows
AttributeError: module 'os' has no attribute 'getuid'
```

The production comment is true for real Windows runtime; these tests *simulate* macOS on Windows, so they do reach `getuid`.

**After** (`monkeypatch.setattr(gateway_mod.os, "getuid", lambda: 501, raising=False)`):

```
11 passed in 0.64s
```

`raising=False` is required here — on Win32 the attribute does not exist, so a strict setattr would fail before the UID stub is installed. Matching `gui/501` is the right value.

Two-line test-only change; no production behavior change. I did not Approve (CI shows no checks on `fix/98038-launchd-test-getuid`).

PR #98068, opened 2026-08-29
fix(computer-use): stop elevated Windows overlays from pinning the desktop

### LINKED ISSUE #98052: [Bug][Windows]: cua-driver overlay survives Hermes shutdown — fullscreen, unclosable without elevation
## Summary

On Windows, after a computer-use session, the cua-driver cursor overlay (a fullscreen, always-on-top layered window, class `Cua.AgentCursorOverlay`) can remain on screen after the Hermes desktop app is closed. It has no close affordance, survives Hermes restarts and new sessions/chats, and cannot be dismissed or killed from a non-elevated shell — the daemon runs elevated (autostart Scheduled Task, RunLevel=Highest), so `taskkill /F` returns Access Denied and `ShowWindow(SW_HIDE)` / `WM_CLOSE` from a normal-integrity process are blocked by UIPI.

From the user's perspective this is "an unclosable window stuck on my screen" — they end up rebooting or living with it.

## Environment

- Windows 11 (Terracorp), Hermes Agent v0.20.6 (2026.8.27, git install), desktop app
- cua-driver 0.19.2 at `%LOCALAPPDATA%\Programs\Cua\cua-driver\bin\cua-driver.exe`
- Autostart Scheduled Task: `\cua-driver-serve` — at interactive logon, RunLevel=Highest, task action launches `cua-driver.exe serve` (no `--no-overlay`), last run 8/29/2026 9:01 AM
- `computer_use.no_overlay` unset (Windows default: overlay on)

## Steps to Reproduce

1. Windows 11 with cua-driver autostart enabled; daemon running elevated from the logon task.
2. Run a computer-use session from the Hermes desktop app (overlay enabled).
3. Quit Hermes entirely (close the app) at/after the session — e.g. user shuts the app down when a turn misbehaves.
4. The overlay window stays on screen: class `Cua.AgentCursorOverlay.default`, rect 0,0→1280,720 (fullscreen), `WS_EX_LAYERED|WS_EX_TOPMOST` (EXSTYLE 0x08080020).
5. Relaunching Hermes, starting new chats, even repeated app restarts do not clear it.

## Diagnostics collected

- Overlay: hwnd 131228, fullscreen 1280×720, topmost layered.
- Daemon: pid 13316, parent pid 8260 already dead (orphaned); `%LOCALAPPDATA%\cua-driver\cua-driver.pid` still names it.
- Non-elevated attempts: `taskkill /PID 13316 /F` → "Access is denied"; `ShowWindow(SW_HIDE)` → returns False; `WM_CLOSE` → ignored.
- Elevated helper: `ShowWindow(SW_HIDE)` → True (overlay gone), `Stop-Process -Force` kills the daemon; final `cua-driver` process count 0.

## Root-cause hypothesis

`cua-driver serve --help` says each MCP session's cursor "is removed when the session ends." When Hermes exits without a clean session teardown, the elevated daemon is orphaned and nothing ever clears that session's overlay. On Windows the overlay is a fullscreen always-on-top window, so the orphaned session leaves an unclosable layer over the entire desktop. This is the same failure shape that motivated defaulting `--no-overlay` on X11 (fullscreen, always-on-top, all-workspaces per `tests/computer_use/test_cua_no_overlay.py`); Windows has the identical window geometry but keeps the overlay on.

## Suggested fixes

1. **Daemon-side watchdog:** when the owning session/client is gone, tear down that session's overlay — extends the coverage of `tests/computer_use/test_cua_atexit_teardown.py` to the orphaned-daemon case on Windows.
2. **Default `no_overlay` on Windows** (as on X11), or at minimum when the daemon was started by the autostart Scheduled Task rather than an interactive spawn.
3. **Recovery path:** a documented, user-runnable way to clear a stuck overlay (e.g. `cua-driver stop` reaching the elevated daemon, or a doctor check that detects and hides orphaned overlay windows).

## Workaround applied locally

- `hermes config set computer_use.no_overlay true`
- Re-registered `\cua-driver-serve` with `serve --no-overlay`

fangliquanflq:
## What does this PR do?

Windows users can be left with a fullscreen, always-on-top cursor overlay after Hermes exits when cua-driver was started by an elevated autostart task. A normal-integrity Hermes process cannot hide or terminate that elevated window, so this change disables the cosmetic overlay by default on Windows while preserving an explicit opt-in.

### Symptom

After a computer-use session and Hermes shutdown, `Cua.AgentCursorOverlay.default` can remain fullscreen and always on top. Relaunching Hermes does not clear it, and non-elevated `WM_CLOSE`, `ShowWindow`, and process termination attempts are rejected by the Windows integrity boundary.

### Impact

Affected Windows users can have the desktop obscured by a window with no close affordance until an elevated recovery action or reboot removes the elevated cua-driver process.

### Bug Cause

**Trigger:** `tools/computer_use/cua_backend.py:233` / `_cua_no_overlay()` returning `False` for the default Windows configuration.

**Causal chain:**
1. An elevated scheduled task starts the shared cua-driver runtime with cursor overlays enabled.
2. Hermes starts a computer-use session, then exits before the remote runtime reliably tears down that session's overlay.
3. The elevated fullscreen overlay survives, while the normal-integrity Hermes process is blocked by Windows UIPI from dismissing it.

**Why it is wrong:** The overlay is cosmetic, but the default enabled it on a platform where the client process cannot reliably recover an orphan owned by a higher-integrity runtime.

**Working sibling / contrast:** macOS and Linux X11 already default the overlay off for equivalent persistent-overlay failure modes while keeping computer-use input and capture functionality intact.

**Ruled out:** Killing or hiding the elevated daemon from Hermes is not a safe recovery mechanism. The captured non-elevated process and window operations fail with access denied, and adding an elevation path would expand privilege rather than remove the failure source.

### Fix

Treat Windows like the existing macOS and Linux X11 safe defaults: auto-detected policy now passes `--no-overlay` and disables the session cursor. Users who accept the lifecycle risk can still set `computer_use.no_overlay: false` to restore it. The config comments and computer-use documentation now describe the Windows behavior.

## Related Issue

Closes #98052

## Type of Change

## Changes Made

- `tools/computer_use/cua_backend.py` - default the cosmetic cursor overlay off on Windows.
- `tests/computer_use/test_cua_no_overlay.py` - cover the Windows default and explicit opt-in on a native Windows test host.
- `hermes_cli/config_defaults.py` - document the auto-detected Windows policy.
- `website/docs/user-guide/features/computer-use.md` - explain the Windows safety default and opt-in.

## How to Test

1. Run the computer-use overlay policy and teardown tests on Windows:

```bash
scripts/run_tests.sh tests/computer_use/test_cua_atexit_teardown.py tests/computer_use/test_cua_no_overlay.py -q
```

2. Confirm the result reports 14 passed and 0 failed on Windows, including the native `windows_only` policy cases.

### Code

- [x] I've run the repository test entry on the relevant tests and all tests pass
- [x] I've tested the policy behavior on Windows 11

- [x] I've updated relevant documentation
- [x] Config example update is N/A because no config key was added or changed
- [x] Architecture/workflow documentation update is N/A
- [x] I've considered cross-platform impact; explicit overrides and Linux/macOS behavior remain unchanged
- [x] Tool schema updates are N/A because the model-facing tool contract is unchanged

## Screenshots / Logs

The targeted Windows test run completed with 14 passed, 0 failed, and 6 host-gated Linux/macOS cases skipped.

### DIFF
--- a/hermes_cli/config_defaults.py
+++ b/hermes_cli/config_defaults.py
@@ -3720,10 +3720,11 @@ DEFAULT_CONFIG = {
         "capture_after_mode": "som",
         # Disable the cursor overlay rendered by cua-driver. The overlay
         # shows where agent actions land but can peg a core when idle
-        # (macOS vImage redraw loop #47032; Linux/WSL2 idle spin #28152).
+        # (macOS vImage redraw loop #47032; Linux/WSL2 idle spin #28152;
+        # elevated Windows daemon orphan overlay #98052).
         # cua-driver ≥ 0.6.x supports --no-overlay; Hermes also calls
         # set_agent_cursor_enabled(false) after start_session when this is on.
-        #   None  = auto-detect (off on macOS + headless/WSL2 Linux; on elsewhere)
+        #   None  = auto-detect (off on macOS, Windows, and headless/WSL2/X11 Linux)
         #   True  = always disable the overlay
         #   False = always enable the overlay
         "no_overlay": None,
--- a/tests/computer_use/test_cua_no_overlay.py
+++ b/tests/computer_use/test_cua_no_overlay.py
@@ -4,7 +4,8 @@ cua-driver's cursor overlay rendering loop can consume CPU indefinitely when
 idle (#28152, #47032), and on Linux/X11 its fullscreen always-on-top overlay
 window can wedge the desktop when a session ends uncleanly. Hermes passes
 ``--no-overlay`` to suppress it when the ``computer_use.no_overlay`` config is
-enabled (or auto-detected on macOS, headless Linux / WSL2, and Linux X11).
+enabled (or auto-detected on macOS, Windows, headless Linux / WSL2, and Linux
+X11).
 
 These assert the behavior contract (auto-detect, explicit override, version
 probe), not specific config snapshots.
@@ -28,6 +29,19 @@ class TestNoOverlayFlag:
         with patch("hermes_cli.config.load_config",
                    return_value={"computer_use": {"no_overlay": True}}):
             assert cua_backend._cua_no_overlay() is True
+
+    @pytest.mark.windows_only
+    def test_windows_auto_detects_off(self):
+        """Windows defaults off so an elevated orphan cannot pin the desktop."""
+        with patch("hermes_cli.config.load_config", return_value={}):
+            assert cua_backend._cua_no_overlay() is True
+
+    @pytest.mark.windows_only
+    def test_windows_explicit_false_restores_cursor(self):
+        """Users may explicitly opt back into the cosmetic Windows cursor."""
+        with patch("hermes_cli.config.load_config",
+                   return_value={"computer_use": {"no_overlay": False}}):
+            assert cua_backend._cua_no_overlay() is False
 
 
     @pytest.mark.macos_only
--- a/tools/computer_use/cua_backend.py
+++ b/tools/computer_use/cua_backend.py
@@ -236,18 +236,21 @@ def _cua_no_overlay() -> bool:
     Reads ``computer_use.no_overlay``. Default ``None`` (auto-detect):
     disable the overlay where idle CPU burn or an X11 desktop wedge is a
     known failure mode — macOS (cursor-overlay vImage redraw loop,
-    #28152/#47032), headless Linux / WSL2 / containers, and Linux X11
-    (fullscreen always-on-top overlay window that can get stuck over every
-    workspace after an unclean session end) — and keep it on Windows and
-    Linux Wayland. Explicit ``True`` / ``False`` overrides auto-detection.
+    #28152/#47032), Windows (an elevated scheduled-task daemon can leave an
+    overlay that a normal-integrity Hermes process cannot dismiss), headless
+    Linux / WSL2 / containers, and Linux X11 (fullscreen always-on-top overlay
+    window that can get stuck over every workspace after an unclean session
+    end) — and keep it on Linux Wayland. Explicit ``True`` / ``False``
+    overrides auto-detection.
     """
     val = _computer_use_cfg().get("no_overlay")
     if val is not None:
         return bool(val)
-    # Auto-detect: macOS overlay can peg a core indefinitely after a
-    # computer_use session (#47032). Prefer off until the driver teardown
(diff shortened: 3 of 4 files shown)


Enough1122:

**Overall:** Disables the cosmetic `cua-driver` cursor overlay on Windows by default to avoid an orphaned fullscreen always-on-top window.

**What it does**
- `tools/computer_use/cua_backend.py:78` extends auto-detect from `darwin` only to `{"darwin","win32"}` — `computer_use.no_overlay: None` now returns `True` on Windows (`_cua_no_overlay`). Explicit `true`/`false` in `config.yaml` still overrides.
- `hermes_cli/config_defaults.py:15` and `website/docs/user-guide/features/computer-use.md:95` update the `None = auto-detect` doc to include Windows (#98052).
- Tests `tests/computer_use/test_cua_no_overlay.py:38` add `windows_only` cases for both auto-off and explicit opt-in (`False` restores cursor).

**Non-blocking notes**
- Behavior change is intentionally fleet-wide for Windows hosts running an elevated autostart daemon — users who want the cursor back must set `computer_use.no_overlay: false`; the PR description and docs make that clear.
- No change to `Linux Wayland` (still on) vs `Linux X11`/`headless`/`WSL2` (off) — correct.

No functional issues found.

fangliquanflq:
Thanks for the review. I checked each point against head `5a2eddcf77620f251347b7e31f73ad52cdb21bc8` and current `main`:

- `_cua_no_overlay()` now defaults to `True` on native Windows while preserving explicit `true`/`false` overrides.
- The config comment and computer-use documentation describe the same Windows default and the `computer_use.no_overlay: false` opt-in.
- The Windows-only behavior tests cover both the safe default and explicit opt-in. The Linux X11/headless/WSL2 and Wayland branches are unchanged by this diff.
- The existing launch and session paths continue to apply the policy via `--no-overlay` when supported and `set_agent_cursor_enabled(false)` after session start.

No follow-up code change is needed. I reran `scripts/run_tests.sh tests/computer_use/test_cua_atexit_teardown.py tests/computer_use/test_cua_no_overlay.py -q` on Windows: **14 passed, 0 failed, 6 host-gated Linux/macOS tests skipped**. The required Windows CI check is also green.

PR #98069, opened 2026-08-29
Keep inherited profile credentials owned by the global store

Alvarado1-ai:
A named profile that inherits a global credential currently can persist runtime status/refresh changes into its profile auth store, materializing a local shadow. This change tracks the owning persistence path: inherited runtime writes go to the global owner, explicit profile add starts an empty local pool, and profile-scoped removal raises a distinct ownership refusal.

It also provides read-only canary suppression for persistence/recovery and exercises existing cross-process flock plus atomic fsync/replace behavior.

Proof: expanded auth/pool suites 139/139, including two-process merge, crash-before-rename, 401 global-owner persistence, no local shadow, and explicit local add isolation.

### DIFF
--- a/agent/credential_pool.py
+++ b/agent/credential_pool.py
@@ -75,6 +75,20 @@ STATUS_EXHAUSTED = "exhausted"
 # login) rewrites the tokens.
 STATUS_DEAD = "dead"
 
+
+class CredentialOwnershipRefused(RuntimeError):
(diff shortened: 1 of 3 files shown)


alt-glitch:

Duplicate of #90992. Both PRs preserve the owner store for profile-inherited credential-pool runtime writes; #90992 is the earlier, broader source-aware implementation with concurrent refresh protection.

PR #96435, opened 2026-08-27
test(cli): isolate update HEAD-movement tests from live runtimes

HermesPrimeHQ:
The HEAD-movement tests simulate a successful update but did not use the shared runtime-isolation fixture. The simulated update could discard gateway mocks during module refresh and enter real runtime recovery paths.

Apply the existing `isolated_update_runtime` fixture to this test module. It keeps the stale-module purge, process discovery, desktop cleanup, and fleet-restart boundaries stubbed using their current namespaces. Upstream already covers the autostash and `--yes` paths, so their old changes and obsolete namespace-introspection tests are no longer needed in this PR.

Validation: the successful HEAD-movement case failed on unmodified upstream c3ce41645c; after the fixture change, all 38 tests across the HEAD-movement, autostash, and `--yes` files passed with `scripts/run_tests.sh` on Linux. No host-OS impersonation is used.

PR #98070, opened 2026-08-29
test(update): keep updater fixtures isolated from live gateways

### LINKED ISSUE #98037: [Bug]: updater tests can reach live gateways after module purge invalidates their mocks
## Bug description

Two updater tests that explicitly claim to use no live gateway can rediscover the developer's real Hermes gateway processes after `hermes update` purges and reimports `hermes_cli.gateway`:

- `tests/hermes_cli/test_update_head_moved_gate.py::test_update_success_when_head_moves`
- `tests/hermes_cli/test_update_fleet_restart_pending.py::test_marker_written_after_pull_cleared_after_successful_restart`

The repository's live-system guard prevented any signal from reaching those processes, so this currently fails safely. The tests are nevertheless non-hermetic and cannot validate their intended updater contracts on a development machine with a gateway running.

## Reproduction

Environment: Windows 11 build 26200, Python 3.11.7, pytest 9.1.1. Tested on `main@3f36c87e1e`; the implicated files are unchanged through current `main@835a913ffd`.

With at least one real Hermes gateway running:

```powershell
.\.venv\Scripts\python.exe -m pytest -q --tb=short `
  tests/hermes_cli/test_update_head_moved_gate.py::test_update_success_when_head_moves `
  tests/hermes_cli/test_update_fleet_restart_pending.py::test_marker_written_after_pull_cleared_after_successful_restart
```

Both cases fail after the guard blocks `os.kill(<real gateway pid>, 15)` because the PID is outside the test process subtree.

## Root cause

Both fixtures patch `hermes_cli.gateway.find_gateway_pids` before invoking the updater. The update path then calls `_purge_stale_hermes_modules()` at `hermes_cli/update_cmd.py:8896` and performs a function-local import of `find_gateway_pids` from a newly loaded `hermes_cli.gateway` at line 8898. The monkeypatch remains attached to the discarded module object, so the new import uses the real process scanner.

This control demonstrates the boundary without changing product behavior:

```python
monkeypatch.setattr(hermes_main, "_purge_stale_hermes_modules", lambda: None)
```

Adding that test-only control to both fixtures changes the same two-node run from RED to `2 passed in 3.86s`.

## Expected behavior / closure criteria

- These tests never discover, signal, stop, or restart a Hermes process outside their own test process tree.
- Both cases pass with a real gateway running and with no gateway running.
- The production stale-module purge remains covered by tests that exercise that behavior intentionally; this fix must not weaken the updater's purge.
- A negative control proves the live-system guard still rejects an intentional attempt to signal an unrelated process.

## Related work (not duplicates)

- #79678 / #86687 introduced the HEAD-movement behavior and its original fixture. They address updater correctness, not the fixture becoming invalid after a module purge.
- #95294 owns interrupted-update fleet reconciliation. It does not own test-process isolation.
- #97633 / #97634 are a behavior-preserving decomposition. The same failure exists on clean `main`, so it is not caused by that refactor.

I found no existing issue or active PR for this specific fixture/reimport boundary.

sylvesterkaczmarek:
## Summary

- keep the HEAD-movement and fleet-restart updater fixtures hermetic across the update flow
- prevent their mocked gateway discovery from being invalidated by the stale-module purge
- use production-compatible gateway scanner mocks and patch restart helpers on their actual `hermes_cli.update_cmd` bindings
- leave production module-purge and gateway behavior unchanged

## Root cause

Both fixtures patch `hermes_cli.gateway` so gateway discovery returns no live processes. During a successful update, the production path deliberately calls `_purge_stale_hermes_modules()` and later reimports `hermes_cli.gateway`.

That creates a new module object, while the fixture's monkeypatches remain attached to the discarded one. The tests can therefore fall through to the real gateway process scanner and attempt to interact with a developer's running Hermes gateway.

The stale-module purge is not part of either test's contract. Each shared fixture patches `hermes_main._purge_stale_hermes_modules` to a no-op, keeping the gateway mocks attached for the duration of the test.

The fixtures also need to match the production restart surface: `find_gateway_pids` can receive additional keyword arguments such as `exclude_pids`, while `_finish_dashboard_update_cleanup` and `_restart_macos_launchd_gateways` are invoked through bindings in `hermes_cli.update_cmd`.

## Changes

- `tests/hermes_cli/test_update_head_moved_gate.py`
  - preserve gateway mocks across the separately tested stale-module purge
  - accept the production PID-scanner call signature
  - stub the dashboard cleanup and macOS launchd restart helpers on `hermes_cli.update_cmd`
- `tests/hermes_cli/test_update_fleet_restart_pending.py`
  - apply the same isolation to the fleet-restart fixture

No production code is changed.

## Testing

Reviewer validation on exact current head `10f215cf29e67e823cd2394b7d8bebbf814db119` under Windows 11 Home build 26200, Python 3.11.7 and pytest 9.1.1:

- `tests/hermes_cli/test_update_head_moved_gate.py`
- `tests/hermes_cli/test_update_fleet_restart_pending.py`
- `tests/hermes_cli/test_update_stale_module_purge.py`
- result: `20 passed in 11.81s`

This confirms the two modified fixtures pass with the production-compatible PID mock and corrected `hermes_cli.update_cmd` bindings, while the dedicated stale-module-purge regression suite remains green.

## Overlap

#96435 partially overlaps the `test_update_head_moved_gate.py` isolation and is still open, but it does not cover `test_update_fleet_restart_pending.py`. This PR remains independently complete for #98037; if the overlapping change lands first, the duplicated HEAD-movement fixture change can be dropped during rebase.

Closes #98037

### DIFF
--- a/tests/hermes_cli/test_update_fleet_restart_pending.py
+++ b/tests/hermes_cli/test_update_fleet_restart_pending.py
@@ -97,6 +97,10 @@ def _patch_update_deps(monkeypatch, tmp_path, run_side_effect):
     monkeypatch.setattr(
         hermes_main, "_resume_windows_gateways_after_update", lambda *a, **k: None
     )
+    # This fixture validates fleet-restart bookkeeping, not updater
+    # module-generation refresh. Keep gateway mocks attached to the configured
+    # module object so the test cannot rediscover live gateway processes.
+    monkeypatch.setattr(hermes_main, "_purge_stale_hermes_modules", lambda: None)
     monkeypatch.setattr(hermes_main, "_write_update_incomplete_marker", lambda: None)
     monkeypatch.setattr(hermes_main, "_clear_update_incomplete_marker", lambda: None)
     monkeypatch.setattr(
--- a/tests/hermes_cli/test_update_fleet_restart_pending.py
+++ b/tests/hermes_cli/test_update_fleet_restart_pending.py
@@ -103,8 +103,13 @@ def _patch_update_deps(monkeypatch, tmp_path, run_side_effect):
     monkeypatch.setattr(hermes_main, "_purge_stale_hermes_modules", lambda: None)
     monkeypatch.setattr(hermes_main, "_write_update_incomplete_marker", lambda: None)
     monkeypatch.setattr(hermes_main, "_clear_update_incomplete_marker", lambda: None)
+    # These restart helpers are bare globals in hermes_cli.update_cmd, so
+    # patch that namespace rather than the hermes_cli.main re-export surface.
     monkeypatch.setattr(
-        hermes_main, "_finish_dashboard_update_cleanup", lambda *a, **k: None
+        update_cmd, "_finish_dashboard_update_cleanup", lambda *a, **k: None
+    )
+    monkeypatch.setattr(
+        update_cmd, "_restart_macos_launchd_gateways", lambda *a, **k: None
     )
     monkeypatch.setattr(hermes_main, "_build_web_ui", lambda *a, **k: None)
     monkeypatch.setattr(
@@ -115,7 +120,7 @@ def _patch_update_deps(monkeypatch, tmp_path, run_side_effect):
     import hermes_cli.gateway as hermes_gateway
 
     monkeypatch.setattr(
-        hermes_gateway, "find_gateway_pids", lambda all_profiles=False: []
+        hermes_gateway, "find_gateway_pids", lambda *a, **k: []
     )
     monkeypatch.setattr(hermes_gateway, "supports_systemd_services", lambda: False)
     monkeypatch.setattr(
--- a/tests/hermes_cli/test_update_head_moved_gate.py
+++ b/tests/hermes_cli/test_update_head_moved_gate.py
@@ -107,15 +107,23 @@ def _patch_update_deps(monkeypatch, tmp_path, run_side_effect):
     # Short-circuit the long tail: dependency install + desktop build.
     monkeypatch.setattr(hermes_main, "_write_update_incomplete_marker", lambda: None)
     monkeypatch.setattr(hermes_main, "_clear_update_incomplete_marker", lambda: None)
-    # Gateway restart path (called after a successful update).
-    monkeypatch.setattr(hermes_main, "_finish_dashboard_update_cleanup", lambda *a: None)
+    # These restart helpers are bare globals in hermes_cli.update_cmd, so
+    # patch that namespace rather than the hermes_cli.main re-export surface.
+    monkeypatch.setattr(
+        "hermes_cli.update_cmd._finish_dashboard_update_cleanup",
+        lambda *a, **k: None,
+    )
+    monkeypatch.setattr(
+        "hermes_cli.update_cmd._restart_macos_launchd_gateways",
+        lambda *a, **k: None,
+    )
     # Keep the (now surfaced — #78574) gateway auto-restart phase away from
     # this machine's real gateways: discovery returns nothing, systemd is
     # unsupported, so the phase is a clean no-op for both snapshots.
     import hermes_cli.gateway as hermes_gateway
 
     monkeypatch.setattr(
-        hermes_gateway, "find_gateway_pids", lambda all_profiles=False: []
+        hermes_gateway, "find_gateway_pids", lambda *a, **k: []
     )
     monkeypatch.setattr(
         hermes_gateway, "supports_systemd_services", lambda: False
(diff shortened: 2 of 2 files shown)


cervantesh:
Thanks for isolating these tests from the production module-purge behavior.

I reproduced the live-system escape on current `main`: the two targeted tests reached the guard while attempting to signal an unrelated gateway process. On this PR, both targeted tests pass, as do all 15 tests in the modified files. The production stale-module purge remains intact, and its dedicated regression tests also pass.

I found two fixture details worth addressing before merge:

1. Both fixtures replace `find_gateway_pids` with `lambda all_profiles=False: []`, while the production call also supplies `exclude_pids=...`. Exercising that exact call currently raises `TypeError`. Using a replacement that accepts the production signature, such as `*args, **kwargs`, would make the isolation explicit instead of depending on surrounding error handling.

2. The macOS restart helper is called through `hermes_cli.update_cmd._restart_macos_launchd_gateways`, but these fixtures do not replace that binding. This leaves the real LaunchAgent path reachable when the platform branch is macOS. Similarly, `_finish_dashboard_update_cleanup` is currently patched on `hermes_cli.main`, while the updater uses its binding in `hermes_cli.update_cmd`.

There is also partial overlap with #96435. That PR already isolates `test_update_head_moved_gate.py` and uses the production-compatible PID-scanner signature and the `update_cmd` bindings, but it does not cover `test_update_fleet_restart_pending.py`. It may therefore be simplest to coordinate the shared `head_moved` change there and retain this PR’s independent fleet-restart coverage, or rebase and remove whichever duplicate lands first.

sylvesterkaczmarek:
> Thanks for isolating these tests from the production module-purge behavior.
> 
> I reproduced the live-system escape on current `main`: the two targeted tests reached the guard while attempting to signal an unrelated gateway process. On this PR, both targeted tests pass, as do all 15 tests in the modified files. The production stale-module purge remains intact, and its dedicated regression tests also pass.
> 
> I found two fixture details worth addressing before merge:
> 
> 1. Both fixtures replace `find_gateway_pids` with `lambda all_profiles=False: []`, while the production call also supplies `exclude_pids=...`. Exercising that exact call currently raises `TypeError`. Using a replacement that accepts the production signature, such as `*args, **kwargs`, would make the isolation explicit instead of depending on surrounding error handling.
> 2. The macOS restart helper is called through `hermes_cli.update_cmd._restart_macos_launchd_gateways`, but these fixtures do not replace that binding. This leaves the real LaunchAgent path reachable when the platform branch is macOS. Similarly, `_finish_dashboard_update_cleanup` is currently patched on `hermes_cli.main`, while the updater uses its binding in `hermes_cli.update_cmd`.
> 
> There is also partial overlap with #96435. That PR already isolates `test_update_head_moved_gate.py` and uses the production-compatible PID-scanner signature and the `update_cmd` bindings, but it does not cover `test_update_fleet_restart_pending.py`. It may therefore be simplest to coordinate the shared `head_moved` change there and retain this PR’s independent fleet-restart coverage, or rebase and remove whichever duplicate lands first.


Thanks, agreed on all three points.

I've updated both fixtures so `find_gateway_pids` accepts the production call shape, moved `_finish_dashboard_update_cleanup` to the `hermes_cli.update_cmd` binding, and also stubbed `_restart_macos_launchd_gateways` there.

I've kept the `head_moved` change in this PR for now so #98037 remains independently covered across both affected tests. I also noted the partial overlap with #96435 in the PR description; if that lands first, I'll drop the duplicated HEAD-movement fixture change during rebase.

I haven't rerun the Windows validation after these fixture-only corrections.

cervantesh:
Thanks for addressing the review points. I reran the focused validation on the updated exact head `10f215cf29e67e823cd2394b7d8bebbf814db119`:

- `tests/hermes_cli/test_update_head_moved_gate.py`
- `tests/hermes_cli/test_update_fleet_restart_pending.py`
- `tests/hermes_cli/test_update_stale_module_purge.py`

Result: **20 passed in 11.81s**.

Environment: Windows 11 Home build 26200, Python 3.11.7, pytest 9.1.1.

The two modified fixture suites pass with the production-compatible PID mock and the corrected `hermes_cli.update_cmd` bindings. The dedicated stale-module-purge regression suite also remains green, so this focused run found no regression in the production purge behavior.

sylvesterkaczmarek:
> Thanks for addressing the review points. I reran the focused validation on the updated exact head `10f215cf29e67e823cd2394b7d8bebbf814db119`:
> 
> * `tests/hermes_cli/test_update_head_moved_gate.py`
> * `tests/hermes_cli/test_update_fleet_restart_pending.py`
> * `tests/hermes_cli/test_update_stale_module_purge.py`
> 
> Result: **20 passed in 11.81s**.
> 
> Environment: Windows 11 Home build 26200, Python 3.11.7, pytest 9.1.1.
> 
> The two modified fixture suites pass with the production-compatible PID mock and the corrected `hermes_cli.update_cmd` bindings. The dedicated stale-module-purge regression suite also remains green, so this focused run found no regression in the production purge behavior.

Thanks for rerunning it on the updated head. I've added the 20/20 focused result to the PR description.

PR #94891, opened 2026-08-25
fix(mcp): honor SDK snake-case read-only annotations

MRPANACHE:
## Summary
- recognize MCP SDK 2.x ToolAnnotations.read_only_hint alongside legacy readOnlyHint
- preserve fail-closed behavior for missing and non-boolean hints
- add a behavior test using a real mcp.types.Tool parsed from camelCase wire data

## Why
MCP SDK 2.x maps wire readOnlyHint to the Python attribute read_only_hint. Hermes only checked readOnlyHint, so read-only tools on trust: untrusted servers were incorrectly routed through dangerous-command approval.

## Test
TMPDIR=<task scratch> HERMES_PYTHON=<task venv>/bin/python scripts/run_tests.sh tests/tools/test_mcp_trust_gating.py -q

Result: 12 passed.

PR #98074, opened 2026-08-29
fix(mcp): read SDK snake-case safety annotations

hermeschay:
## Summary
- recognize the MCP Python SDK's snake_case `ToolAnnotations.read_only_hint` attribute
- preserve camelCase and cached-dict compatibility
- add a regression test using the real SDK `ToolAnnotations` model

## Why
MCP annotations arrive on the wire as `readOnlyHint`, but the Python SDK exposes that field as `read_only_hint`. Hermes checked only the camelCase object attribute, so live MCP discovery classified every SDK tool as write-capable. On an `untrusted` server this caused read-only calls to require approval and wrote false hints into the lazy schema cache.

## Verification
- `uv run --extra dev pytest -q tests/tools/test_mcp_trust_gating.py tests/tools/test_mcp_schema_cache.py` — 23 passed
- `uv run --extra dev pytest -q tests/tools/test_mcp*.py` — 597 passed
- `uv run --extra dev ruff check tools/mcp_tool.py tests/tools/test_mcp_trust_gating.py` — passed
- reproduced against a live MCP server: `ToolAnnotations.model_dump(by_alias=True)` returned `readOnlyHint: true` while the Python attribute was `read_only_hint=True`

### DIFF
--- a/tests/tools/test_mcp_trust_gating.py
+++ b/tests/tools/test_mcp_trust_gating.py
@@ -231,6 +231,17 @@ class TestAnnotationCaptureAtDiscovery:
         assert not hints.get("delete_repo")
         assert not hints.get("no_annotations")
 
+    def test_sdk_tool_annotations_supported(self):
+        """The MCP SDK exposes aliased hints as snake_case attributes."""
+        from mcp.types import ToolAnnotations
+
+        assert mcp_tool._annotation_read_only_hint(
+            SimpleNamespace(annotations=ToolAnnotations(read_only_hint=True))
+        ) is True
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #94891: both repair `_annotation_read_only_hint()` for the MCP SDK's snake-case `read_only_hint` while retaining camelCase and cached-dict compatibility.

hermeschay:
Closing in favor of the earlier #94891, which addresses the same MCP SDK snake-case annotation compatibility issue. Thanks for the duplicate triage.

PR #98079, opened 2026-08-29
feat(desktop-web): desktop web client implemented

Elshad:
Have some bugs. Actually created PR in this stage because of maybe you will like this idea and support it. 

Screenshot 2026-08-29 at 23 10 23


LustHK:
Interesting; it will probably be much more optimized than the embarrassing Windows desktop app for Hermes Agent. It’s incredibly slow and laggy.

Issue #97294, opened 2026-08-28
[Feature]: Add ZCode and WorkBuddy AI as providers under the account


Skauyy:
### Problem or Use Case

Hi team,

I'd like to ask whether it would be possible to add **ZCode** and **WorkBuddy AI** as providers under the account settings, so I can use them directly in Hermes.

Thanks!

### Proposed Solution

Add **ZCode** and **WorkBuddy AI** as providers under the account settings (e.g. Account → Providers).

- In **Account → Providers**, users should be able to add ZCode and WorkBuddy AI alongside the existing providers.
- After selecting one, the user could log in with their account via OAuth.
- Once added, they should be available for selection and work the same way as the other supported providers.

### Feature Type

Other

### Scope

None

### Contribution

- [ ] I'd like to implement this myself and submit a PR

### Debug Report (optional)

```shell

```

PR #60686, opened 2026-07-08
feat(aux): force streaming for providers that reject non-stream requests

kudi88:
## Problem

Some OpenAI-compatible endpoints — notably **Tencent Copilot** (`copilot.tencent.com`) — only accept **streaming** chat requests. Any non-streaming call returns HTTP 400:

```json
{"code": 11101, "msg": "Non-stream chat request is currently not supported"}
```

The main conversation loop already uses streaming (`conversation_loop.py` line ~1257: *"Always prefer the streaming path"*), so interactive chat works fine with these providers. However, **auxiliary tasks** (title generation, context compression, vision analysis, web extraction, session search, skills hub, MCP routing) all default to **non-streaming** requests via `call_llm()` in `agent/auxiliary_client.py`, causing them to fail silently:

```
⚠ Auxiliary title generation failed: HTTP 400: {"code":11101,"msg":"Non-stream chat request is currently not supported"}
```

This affects any user who configures a custom OpenAI-compatible provider whose endpoint requires streaming.

### New helper functions

| Function | Purpose |
|---|---|
| `_provider_requires_stream(provider, base_url)` | Detects providers that only accept streaming. Currently matches `copilot.tencent.com`; extensible via provider name suffix `:stream-only` |
| `_consume_stream_to_response(stream, model)` | Consumes an OpenAI SDK streaming iterator and returns a `SimpleNamespace` response with `.choices[0].message.content` — matching the shape `_validate_llm_response` already accepts for adapter responses (line ~6292) |
| `_create_and_validate(client, kwargs, task)` | Sync wrapper: if `stream=True` in kwargs, consumes the stream before validation; otherwise falls through to the original path |
| `_create_and_validate_async(client, kwargs, task)` | Async version of the above |

### Modified code paths

1. **`_build_call_kwargs()`**: Automatically sets `stream=True` in kwargs when `_provider_requires_stream` returns True, so all code paths (including fallback retries that rebuild kwargs) inherit the flag.

2. **`call_llm()` (sync + async)**: Added force-stream check before the try block (using the client's actual `base_url`), and replaced all 17 call sites where `_validate_llm_response(client.chat.completions.create(**kwargs), task)` was used with `_create_and_validate(client, kwargs, task)`.

3. **`_retry_same_provider()` / `_retry_same_provider_async()`**: Same replacement for retry paths.

4. **`_call_fallback_candidate()` / `_call_fallback_candidate_async()`**: Same replacement for fallback paths.

## Transparency

[...]

PR #98080, opened 2026-08-29
feat(providers): add native WorkBuddy provider

owenisas:
Adds WorkBuddy (`www.workbuddy.ai`) as a native provider, same shape as TokenHub / TokenPlan.

Refs #97294.

WorkBuddy is Tencent's consumer assistant. Chat is OpenAI-compatible at `https://www.workbuddy.ai/v2`. Auth is a Bearer token in `WORKBUDDY_ACCESS_TOKEN`. No new adapter.

A custom `base_url` is not enough, because of two real failures I hit live.

Non-stream chat returns HTTP 400 `{"code": 11101, "msg": "Non-stream chat request is currently not supported"}`. The conversation loop already streams, so `hermes chat --provider workbuddy` works. Aux tasks do not stream. Title generation, compression, and vision all fail:

```
⚠ Auxiliary title generation failed: HTTP 400: Error code: 400 - {'code': 11101, 'msg': 'Non-stream chat request is currently not supported'}
```

`_provider_requires_stream()` already treats Tencent Copilot this way (#60686). This PR adds `workbuddy.ai` to that host list so aux calls send `stream=True`.

Without an aux-model mapping, those same tasks also leak to the session default provider. Mapped to `fast-model`.

`/v2/models` 404s, so the catalog is static. Tier slots (`default-model`, `fast-model`, `balanced-model`, `primary-model`, `deep-model`) resolve server-side. Named models come from `/v3/config`.

```
hermes chat --provider workbuddy --model fast-model -q "Reply with exactly: WORKBUDDY_NATIVE_OK"
WORKBUDDY_NATIVE_OK

Session:  20260829_113601_6814ce
Title:    Verify workbuddy native response
```

The aux warning is gone. Session title came from WorkBuddy, not the default provider. Tool calling checked separately (`get_weather` → `{"city": "Paris"}`).

28 tests in `tests/hermes_cli/test_workbuddy_provider.py`. Aliases `workbuddy`, `workbuddy-ai`, `wb` all resolve.

Token comes from browser OAuth against Tencent. Out of scope here. Unofficial helper at `github.com/owenisas/workbuddy-openai`. This PR does not import it.

Tier-slot context lengths are `/v3/config` `maxInputTokens` as of 2026-08-29. They are ceilings. The slots can move.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -183,6 +183,16 @@
 # TOKENPLAN_BASE_URL=https://api.lkeap.cloud.tencent.com/plan/anthropic
 
 # =============================================================================
+# LLM PROVIDER (WorkBuddy — Tencent consumer plan)
+# =============================================================================
+# OpenAI-compatible endpoint fronting the WorkBuddy subscription. Sign in via
+# your browser and export the resulting bearer token; see
+# https://github.com/owenisas/workbuddy-openai (unofficial) for one way to
+# obtain it. Endpoint is stream-only — Hermes handles that automatically.
+# WORKBUDDY_ACCESS_TOKEN=your_token_here
+# WORKBUDDY_BASE_URL=https://www.workbuddy.ai/v2
+
+# =============================================================================
 # TOOL API KEYS
 # =============================================================================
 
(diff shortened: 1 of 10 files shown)


leonid-dalin:
Outside of the comment slop, [d557769](https://github.com/NousResearch/hermes-agent/pull/98080/commits/d5577694619829e61d31e81cdb3f5f8347b5707c) looks fine. Looking forward for the merge

owenisas:
Fixed it, thanks for reviewing.

PR #98082, opened 2026-08-29
feat(tts): openai_compatible streaming provider for self-hosted servers

narigondelsiglo:
## What

Adds an `openai_compatible` `StreamingTTSProvider` for any server exposing an OpenAI-compatible `POST /v1/audio/speech` with `response_format=pcm` — the self-hosted TTS ecosystem: **qwentts.cpp** (Qwen3-TTS), **LocalAI**, **kokoro-fastapi**, **Speaches**, OpenAI proxies, etc.

## Why

The existing `openai` streamer supports `base_url` but gates `available()` on an API key — so a local keyless server can never be selected as the streaming provider, even though its API is identical. Self-hosted voice is a real use case: I'm running qwentts.cpp (Qwen3-TTS 1.7B Q8) on a Vulkan iGPU with sub-second first-audio and wanted Hermes voice mode to speak through it.

## How

- `@register("openai_compatible")` `OpenAICompatibleStreamer`: `requests` streamed body through the existing `_capped` 16 MiB per-sentence guard, `int16` mono PCM at 24 kHz.
- `available()` requires only `tts.openai_compatible.base_url` in config (nothing to auto-detect). `api_key` optional — `Authorization: Bearer` sent only when set.
- `model` (default `tts-1`) / `voice` passthrough; trailing-slash normalization on `base_url`.
- Config: `tts.provider: openai_compatible` or `tts.streaming.provider: openai_compatible`.

## Testing

- Contract tests in `tests/tools/test_tts_streaming.py`: availability gate (base_url required), keyless streaming (asserts **no** auth header, pcm format, stream=True), Bearer + voice passthrough when configured. 32/32 pass.
- **E2E against a live server** (qwentts.cpp, Qwen3-TTS 1.7B Q8_0, Vulkan iGPU): resolver → `stream()` → 6.16s of valid 24kHz int16 PCM in 77 chunks, wall 1.67s, RTF 0.27.

## Doc

Follows the "Adding a new streaming provider" recipe in `docs/streaming-tts.md` (subclass + `@register` + tests). Happy to add the provider to that doc's capability matrix if you tell me which doc table you want updated.

### DIFF
--- a/tests/tools/test_tts_streaming.py
+++ b/tests/tools/test_tts_streaming.py
@@ -155,6 +155,116 @@ def test_openai_streamer_prefers_configured_api_key(monkeypatch):
     assert captured["client"]["api_key"] == "cfg-key"
 
 
+# ── openai_compatible: self-hosted servers with no mandatory key ─────────
+
+
+def test_openai_compatible_available_requires_configured_base_url(monkeypatch):
+    monkeypatch.setattr(
+        ts, "_load_tts_config",
+        lambda: {"openai_compatible": {"base_url": "http://127.0.0.1:8902/v1"}},
+    )
+    assert ts.OpenAICompatibleStreamer.available() is True
+    monkeypatch.setattr(ts, "_load_tts_config", lambda: {"openai_compatible": {}})
+    assert ts.OpenAICompatibleStreamer.available() is False
+    monkeypatch.setattr(ts, "_load_tts_config", lambda: {})
+    assert ts.OpenAICompatibleStreamer.available() is False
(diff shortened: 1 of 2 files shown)


narigondelsiglo:
Reopening this work as #98096 from the maintainer-facing fork with a corrected commit author email (GitHub noreply instead of a personal address — required by the check-attribution CI job) and a filled-out PR description. Code content is identical (rebased as-is onto `main`; 32/32 tests pass). Closing in favor of #98096 — please review there.

PR #44044, opened 2026-06-11
fix(cli): accept configured MCP prefixed toolsets before discovery

thedavidweng:
## What does this PR do?

Fix the same class of false-positive MCP toolset validation across CLI, TUI, and oneshot (`-z`) startup paths.

Current `main` already suppresses the warning for bare MCP server names (`agentmail`, `notion`, etc.), but it still warns for the documented `mcp-<server>` spelling (`mcp-agentmail`, `mcp-notion`, etc.) before MCP discovery registers runtime toolsets. That causes startup warnings even when the user has configured MCP servers correctly in `config.yaml`.

This PR adds shared pre-discovery MCP toolset helpers and applies them consistently to the three startup validators so `mcp-<server>` is accepted when the underlying MCP server is configured (enabled or disabled), while genuinely unknown toolsets still warn.

Fixes #78102

## Related Issue

Related #41625
Related #23997
Related #29532
Related #5279

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Added shared MCP toolset helpers:
  - `hermes_cli/mcp_toolsets.py`
- Updated CLI startup validator to accept configured MCP prefixed toolsets before discovery:
  - `cli.py`
- Updated `-z` / oneshot validator to accept configured MCP prefixed toolsets before discovery:
  - `hermes_cli/oneshot.py`
- Updated TUI toolset loader to accept configured MCP prefixed toolsets before discovery:
  - `tui_gateway/server.py`
- Added regression tests:
  - `tests/cli/test_mcp_toolset_validation.py`
  - `tests/hermes_cli/test_tui_resume_flow.py`
  - `tests/test_tui_gateway_server.py`

## How to Test

1. Configure `mcp_servers` in `~/.hermes/config.yaml` with one or more servers.
2. Use `platform_toolsets` entries such as:
   - `mcp-agentmail`
   - `mcp-notion`
   - `mcp-notion-findit`
   - `mcp-tinyfish`
3. Start Hermes CLI / TUI / oneshot (`hermes -z`) and confirm no false:
   `Warning: Unknown toolsets: mcp-agentmail, mcp-notion, ...`
4. Run tests:
[...]

PR #98088, opened 2026-08-29
fix(mcp): reuse stdio watcher coroutine

Serph91P:
## Summary
- create the stdio child-watcher coroutine exactly once
- reuse that coroutine for awaitability validation and task scheduling
- prevent the discarded probe coroutine from emitting `RuntimeWarning: coroutine ... was never awaited`
- assert in the reconnect regression test that the watcher is constructed once

The MCP-prefixed pre-discovery validation changes were removed because they are already covered more broadly by #44044.

## Verification
- `uv run pytest -q tests/tools/test_mcp_stdio_fastfail_reconnect.py` (2 passed)
- `uv run pytest -q tests/tools/test_mcp*.py` (596 passed)
- `uv run ruff check tools/mcp_tool.py tests/tools/test_mcp_stdio_fastfail_reconnect.py`
- `git diff --check`

### DIFF
--- a/tests/tools/test_mcp_stdio_fastfail_reconnect.py
+++ b/tests/tools/test_mcp_stdio_fastfail_reconnect.py
@@ -110,8 +110,15 @@ def test_midcall_child_exit_signals_reconnect(monkeypatch, tmp_path):
         mcp_tool, "srv-midcall", _hanging_call, children_dead=lambda: False
     )
 
-    async def _watch_children():
-        return  # children die immediately → watcher resolves first
+    watch_calls = {"n": 0}
+
+    def _watch_children():
+        watch_calls["n"] += 1
+
+        async def _child_exit():
+            return  # children die immediately → watcher resolves first
+
+        return _child_exit()
 
     server._watch_stdio_children = _watch_children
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #44044 for the configured `mcp-<server>` pre-discovery validation repair; #44044 is the broader open implementation across startup validators. This PR also carries separate stdio watcher-warning cleanup, which should be preserved if work is consolidated.

Serph91P:
Thanks for flagging this. I removed the duplicate MCP-prefixed validation changes. This PR now contains only the independent stdio watcher coroutine cleanup; #44044 remains the implementation for pre-discovery validation.

teknium1:
Closing with credit — the leak this fixes landed on main via #100920 (fe3e5dcb9ef), carrying @loulanyue's earlier #96044 fix (Aug 27) with a different shape: the probe now uses `inspect.iscoroutinefunction(_watch_children)` instead of creating the coroutine once and reusing it. Same bug, same outcome (0 leaked coroutines, verified live). Thanks @Serph91P for the independent fix.

PR #98089, opened 2026-08-29
fix(telegram): send explicit video dimensions

shrimpapplepro:
## Bug Description

Telegram can render an uploaded MP4 with the wrong preview aspect ratio when `sendVideo` is called without explicit dimensions. In a live reproduction, a 1280×720 H.264/AAC video was displayed close to 4:3 even though the file itself reported a 16:9 display aspect ratio.

No matching open or closed issue/PR was found for this symptom.

## Root Cause

`TelegramAdapter.send_video()` uploaded the file without `width`, `height`, or `supports_streaming`. Telegram therefore had to infer the video geometry during server-side processing, and that inference was incorrect for the reproduced file.

## Fix

- Best-effort probe the first video stream with `ffprobe` off the asyncio event loop.
- Apply MP4 display-matrix rotation so 90°/270° phone videos send swapped display dimensions.
- Pass the resulting display `width` and `height` explicitly to Telegram.
- Set `supports_streaming=True` for native video uploads.
- Preserve the previous behavior when `ffprobe` is unavailable or the file cannot be probed.
- Add regression coverage for Bot API payloads, display rotation, unavailable `ffprobe`, malformed output, and probe timeouts.

## How to Verify

1. Send a 1280×720 MP4 through a `MEDIA:` response on Telegram.
2. Confirm the message renders as a 16:9 native video preview and streams normally.
3. Confirm the adapter sends `width=1280`, `height=720`, and `supports_streaming=True`.

## Test Plan

- [x] Added regression test for this bug
- [x] Existing Telegram gateway tests pass
- [x] Manual verification of the fix on Telegram

Commands run:

```bash
uv run --python 3.11 --with pytest --with pytest-asyncio pytest tests/gateway/test_telegram*.py tests/gateway/test_tts_media_routing.py -q -o 'addopts='
# 607 passed

uv run --python 3.11 --with ruff ruff check plugins/platforms/telegram/adapter.py tests/gateway/test_telegram_media_read_timeout.py
# All checks passed
```

## Risk Assessment

Low — the change is limited to Telegram native video sends. Probing is best-effort, bounded by a five-second timeout, and runs in a worker thread. If probing fails, the adapter still uploads the video with the previous inference behavior while retaining `supports_streaming=True`.

### DIFF
--- a/plugins/platforms/telegram/adapter.py
+++ b/plugins/platforms/telegram/adapter.py
@@ -15,6 +15,8 @@ import logging
 import os
 import html as _html
 import re
+import shutil
+import subprocess
 import threading
 import time
 from contextvars import ContextVar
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Fixes Telegram preview aspect ratio by sending explicit video dimensions.

**What it does**
- `plugins/platforms/telegram/adapter.py:21` adds `_probe_video_dimensions(video_path)` via `ffprobe -select_streams v:0 -show_entries stream=width,height:stream_side_data=rotation -of json` with 5s timeout and `windows_hide_flags()`; swaps `w/h` on quarter-turn rotation (`adapter.py:68`).
- `adapter.py:88` calls it via `asyncio.to_thread` before `send_video` and includes `supports_streaming=True` plus `width`/`height` when probing succeeds (`adapter.py:91`).
- Best-effort: returns `None` without `ffprobe`, on non-zero exit, bad JSON, or `TimeoutExpired`, so media delivery still works.

**Non-blocking notes**
- `ffprobe` is invoked with `capture_output=True` + `text=True, errors="replace"` and `stdin=DEVNULL` — good hygiene; `check=False` avoids raising on non-zero.
- Thread offload (`to_thread`) prevents blocking the event loop for the 5s probe — correct.
- Rotation handling checks only the first side-data entry with a `rotation` field and uses `round(float(rotation)) % 180 == 90` — covers 90/270 and -90; 180 correctly leaves dimensions unchanged.

No functional issues found.

Issue #98077, opened 2026-08-29
[Field report]: state.db physical cross-B-tree corruption under SQLite 3.50.4 WAL; canonical writes continued after malformed detection


Caelier:
## Bug Description

A production Hermes `state.db` became physically corrupted while running Hermes-linked SQLite 3.50.4 in WAL mode with multiple Hermes processes/connections.

The incident initially surfaced as FTS corruption. Offline forensic recovery found structural corruption beyond FTS, including an incompatible/cross-linked canonical table B-tree child, traversal into another logical object's pages, and access-path-dependent row results.

Hermes disabled FTS synchronization and retried canonical writes after the malformed error. Gateway/Discord operation therefore remained substantially functional while the database was structurally degraded.

The database was recovered by offline logical reconstruction, FTS rebuild from canonical content, writer-excluded cutover, and upgrade to SQLite 3.53.1.

This is a distinct forensic field report, not a claim that one precise SQLite race was captured. It overlaps the diagnostic behavior tracked in [#88587](https://github.com/NousResearch/hermes-agent/issues/88587) and [#97940](https://github.com/NousResearch/hermes-agent/issues/97940), while adding concrete page/tree evidence and complete recovery accounting.

## Environment

```text
Platform: Ubuntu 24.04.4 LTS under WSL2
Filesystem: WSL Linux filesystem (not /mnt/c/DrvFS)
Hermes at diagnosis: v0.20.6 (upstream a73b14c4)
Python: 3.11.15
Hermes SQLite runtime at incident: 3.50.4
Journal mode: WAL
Corrupt source bundle: approximately 1.5 GB
Concurrent Hermes activity: gateway, dashboard/profile services, multiple connections

Post-recovery Hermes: v0.20.6, commit 4209d371aa1bb8840ce8447555bdd863a1a96c38
Post-recovery SQLite: 3.53.1
Post-recovery SQLite source id:
2026-05-05 10:34:17 c88b22011a54b4f6fbd149e9f8e4de77658ce58143a1af0e3785e4e6475127e9
Recovered logical DB: 1,122,344,960 bytes
```

The system `sqlite3` CLI is not used here as evidence for Hermes' linked runtime.

### 2026-08-27

Historical rows later affected by damaged traversal were created.

This is **not claimed to be the corruption timestamp**. A row written on August 27 can become unreachable due to a later page/checkpoint inconsistency.

### 2026-08-28

Gateway activity and canonical writes continued.

[...]

PR #98090, opened 2026-08-29
fix(state): stop structurally corrupt databases from accepting writes

### LINKED ISSUE #97940: SQLITE_CORRUPT on canonical B-trees is misdiagnosed as FTS corruption; logs claim 'canonical message rows are preserved' while transcript writes silently fail
## Summary

`SessionDB._is_fts_write_corruption_error()` treats **every** `database disk image is malformed` as FTS-index corruption (`hermes_state.py`, the `is_malformed_db_error(exc)` branch returns `True` unconditionally). When the damage is actually in the **canonical B-trees** (`messages`, `sessions`), the operator gets logs that assert the opposite of reality, and the automatic response (in-place FTS rebuild) is a no-op against the real problem.

## What we saw (production, 2026-08-29)

A ~7.4 GB `state.db` developed structural damage — full `integrity_check` reported 34 pages failing `btreeInitPage() returns error code 11` across a dozen trees, and `SELECT count(*) FROM messages` itself raised `DatabaseError`. The first symptom was a *canonical read* failing:

```
File ".../hermes_state.py", line 7572, in _row
  row = conn.execute(...)
sqlite3.DatabaseError: database disk image is malformed
```

Yet the log narrative said:

```
WARNING hermes_state: state.db write failed with an FTS-corruption error
  (database disk image is malformed) — attempting one-shot in-place FTS rebuild;
  canonical message rows are preserved.
ERROR hermes_state: In-place FTS rebuild failed (database disk image is malformed);
  the database needs the full offline repair path (repair_state_db_schema).
ERROR hermes_state: state.db FTS indexes remain corrupt (database disk image is
  malformed); disabled FTS sync and retrying the canonical write.
  Search temporarily uses LIKE until a later SessionDB open rebuilds the indexes.
```

All three lines are wrong for this failure mode:

* "FTS-corruption error" — the corrupt pages were canonical table/index B-trees;
* "canonical message rows are preserved" — `messages` was unreadable;
* "FTS indexes remain corrupt … retrying the canonical write" — the retried canonical writes kept failing (`Session DB append_message failed: database disk image is malformed` repeatedly), i.e. the gateway kept running while **silently losing new transcript rows**.

The docstring itself acknowledges the shortcut: *"if the damage is actually in a canonical B-tree, the rebuild itself fails and the write propagates"* — the write does propagate, but the operator-facing story never gets corrected, and the gateway continues in a state where every transcript append fails.

## Why it matters

* An operator (or an auto-repair watchdog) reading these logs will rebuild FTS and restart, believing message data is safe, while the real fix (`repair_state_db_schema` / offline `.recover`) is delayed — in our case data loss continued until an external repair.
* We have seen the generic message mislead twice before in other forms (`session storage could not be written… often a full disk` on a 31 %-used disk).

## Suggested fix

`sqlite3` gives no structured way to distinguish the two cases from the exception alone, but a cheap probe does: after catching a malformed error, attempt `SELECT count(*) FROM messages LIMIT 1` (and `sessions`) on a fresh read-only connection.

* Probe succeeds → keep today's behavior (FTS rebuild), the current wording is then justified.
* Probe fails → skip the pointless FTS rebuild, log an explicit **canonical-corruption** error naming `repair_state_db_schema` as the required path, and consider refusing further silent write-retries (or at least escalating the log level) so the "gateway up but nothing persists" state is loud.

Happy to provide the full incident logs privately if useful.

**Environment:** hermes-agent local build ~Aug 2026 (`0096c9b7` + carried commits), runtime SQLite 3.53.1, Ubuntu 24.04. The classifier at current `main` (`hermes_state.py` ~L5411) is unchanged.

fangliquanflq:
## What does this PR do?

Stops a generic SQLite `database disk image is malformed` error from being treated as proof that only FTS is damaged. Unscoped or mixed corruption now fails closed and durably quarantines the affected database generation across handles and restarts, preventing a structurally damaged database from continuing to accept state changes.

### Symptom

After SQLite reported `database disk image is malformed`, Hermes could label the failure as FTS corruption, rebuild or detach FTS, and retry the canonical write. The gateway could therefore remain substantially functional while canonical B-trees were also structurally damaged.

### Impact

A profile with structural `state.db` corruption could continue canonical writes and appear healthy enough to keep serving. This risks adding more state to a database whose canonical traversal is already unreliable.

### Bug Cause

**Trigger:** `hermes_state.py:SessionDB._is_fts_write_corruption_error` and `gateway/session.py:SessionStore._is_fts_corruption_error` accepted the generic malformed-image message as FTS corruption.

**Causal chain:**
1. A canonical or mixed B-tree failure raises the same generic `SQLITE_CORRUPT` text that older SQLite builds also use for damaged FTS shadow pages.
2. Both runtime classifiers treated that text as FTS-only without checking which object failed.
3. Hermes attempted FTS recovery or detached FTS triggers, then retried canonical writes against a database that might have canonical structural damage.

**Why it is wrong:** The generic SQLite message provides no object provenance. A failed FTS rebuild also does not prove canonical tables are healthy.

**Working sibling / contrast:** Newer SQLite FTS failures that explicitly report `fts5: corrupt structure record` already provide FTS provenance and remain eligible for online recovery.

**Ruled out:** Removing generic FTS recovery entirely would regress older SQLite builds where synthetic FTS shadow corruption raises only the generic malformed-image message. The fix preserves that path when `PRAGMA integrity_check` reports only named Hermes FTS5 tables.

### Fix

- Require explicit FTS result-code provenance or a complete, below-cap integrity report whose every defect is scoped to a known Hermes FTS5 table before online FTS recovery.
- Treat mixed, unscoped, failed, healthy, or cap-truncated integrity results as structural corruption and durably quarantine canonical writes across handles and restarts.
- Stop the gateway retry layer from independently treating generic malformed/schema errors as FTS corruption.
- Keep the existing synthetic FTS corruption self-heal path covered and passing.

## Related Issue

Fixes #97940

Addresses #98077

## Type of Change

## Changes Made

- `hermes_state.py` - verify complete FTS provenance and durably quarantine unscoped structural corruption fail-closed.
- `gateway/session.py` - restrict the gateway fallback classifier to explicit FTS corruption.
- `tests/state/test_fts_runtime_rebuild.py` - cover truncated diagnostics, a physical canonical corruption counterexample, cross-handle write refusal, and retained synthetic FTS recovery.
- `tests/gateway/test_session.py` - cover the gateway classifier boundary.

## How to Test

1. Return a cap-sized FTS-looking integrity report and verify it fails closed as potentially truncated.
2. Physically corrupt the canonical `messages` B-tree while a covering index still answers `COUNT(*)`, then verify no FTS rebuild/detach/retry occurs.
3. Attempt a later canonical write from a sibling handle and verify it is rejected before the callback runs.
4. Corrupt a real synthetic FTS shadow table and verify append/search recovery still succeeds.
5. Run:

```bash
scripts/run_tests.sh tests/state/test_fts_runtime_rebuild.py -k 'not foreign_holder_detection_proc_readlink_deleted_wal and not foreign_holder_uninspectable_process_cmdline_fallback' -q
scripts/run_tests.sh tests/gateway/test_session.py -q
```

Result: 92 passed. The two excluded tests emulate Linux by patching host globals and fail on Windows independently of this change. Independent review also ran six related repair, durability, live-writer, and repair-ledger files: 85 passed and 5 skipped.

### Code

- [x] I've run the repository test entry on the relevant test files and all runnable relevant tests pass

- [x] I've updated relevant docstrings; user documentation is N/A
- [x] `cli-config.yaml.example` is N/A because no config keys changed
- [x] `CONTRIBUTING.md` and `AGENTS.md` are N/A because no workflow changed
- [x] I've considered cross-platform impact; the behavior is SQLite-level and host-independent
- [x] Tool descriptions and schemas are N/A because no tool contract changed

## Screenshots / Logs

N/A - regression tests exercise the persistence behavior directly.


andrexibiza (commented):
Reviewed the exact current object `6736c64c9a2f560e3af179d23b8742e207980c4e` against current `main@4209d371aa1bb8840ce8447555bdd863a1a96c38`, the full four-file diff, #98077/#97940, the older #91585 → merged #92523 lineage, and the still-open #96038/#97843/#88604 surfaces. Exact-head hosted receipts are all green: [CI 33270260313](https://github.com/NousResearch/hermes-agent/actions/runs/33270260313), [Docker 33270259965](https://github.com/NousResearch/hermes-agent/actions/runs/33270259965), and [Nix 33270259944](https://github.com/NousResearch/hermes-agent/actions/runs/33270259944); attribution is green inside the CI run.

The central correction is right: `database disk image is malformed` is not proof of FTS-only damage, and moving that generic text out of both recovery classifiers closes the dangerous availability-over-integrity assumption that #98077 finally disproves with physical cross-B-tree evidence. The synthetic FTS self-heal control is also important — this should not become “all corruption means kill FTS recovery.” I found two remaining P1 ownership/provenance gaps before I would treat `Closes #98077` as true.

### P1 — the structural-corruption quarantine is sticky only on one Python object, not on the profile/database that owns the risk

`_structural_corruption_error` is initialized to `None` in every `SessionDB` instance and all of the new refusal checks read only that field. So process A can observe unscoped/mixed corruption and correctly refuse later writes on *its* handle, while an already-open sibling handle, another Hermes process, or a restarted gateway still has no knowledge that this database has been declared structurally unsafe.

That distinction is material here, not theoretical. #98077's production topology explicitly had multiple Hermes processes/connections, and its forensic result is access-path-dependent corruption: one traversal can fail while another path remains writable enough to keep serving. Therefore “the next handle will immediately see the same malformed error” is not a safe containment assumption. A fresh/restarted `SessionDB` can perform a benign canonical write before it ever touches the damaged page.

The current log text also promises more than the mechanism provides: “canonical writes are disabled for this handle until offline repair” is accurate locally, but the issue asks for a sticky **profile-wide structural-degraded state**, and the PR body says a structurally damaged database stops accepting state changes. Today only one object stops.

Required shape: make the quarantine owned by the database/profile generation rather than the discovering handle, and make every writable participant observe it **pre-effect**. The durable signal cannot live only inside the database whose structural trust has just been revoked; it needs a safe external/shared authority (or equivalent cross-process mechanism), bound tightly enough to the database generation that a verified offline repair/cutover can retire the old quarantine without accidentally blessing the old file. Please add at least these regressions:

1. handle A observes generic structural corruption; already-open handle B attempts an otherwise-valid canonical write and is refused before its callback mutates;
2. A closes, the DB is reopened/restarted without repair, and the new writable handle is still refused;
3. a verified repaired/replaced generation can clear/retire the quarantine without making the corrupt generation writable again.

Until that exists, #98077 should stay open even if the same-handle latch lands.

### P1 — #96038 owns stronger SQLite provenance than this branch currently preserves; absorb it or declare merge order

The new `_has_fts_corruption_evidence()` still starts with `_is_fts_write_corruption_error()`, which accepts any exception whose text contains both `fts5` and `corrupt`. It does not consult `sqlite_errorcode`. Open #96038 (`diatche`) already covers the other side of this exact classifier boundary: known SQLite result-code provenance must outrank prose. It accepts `SQLITE_CORRUPT_VTAB` even when older SQLite emits only generic malformed-image text, and it rejects contradictory known codes even when the message looks FTS-shaped.

That matters in both directions:

- a generic-text `SQLITE_CORRUPT_VTAB` is affirmative virtual-table provenance, but this branch can run `integrity_check`, see `ok`/unscoped output, and incorrectly latch the whole DB structural;
- FTS-looking prose paired with a known contradictory result code currently short-circuits straight to online FTS mutation instead of failing closed.

The repair here should compose the two signals, not replace one with the other: known extended result code first (`SQLITE_CORRUPT_VTAB` = FTS evidence; contradictory known code = reject), then the conservative integrity/provenance probe for the generic/no-code case. Please preserve #96038/`diatche` as complementary work or explicitly supersede/absorb that implementation with credit; the two PRs edit the same `hermes_state.py` classifier and `tests/state/test_fts_runtime_rebuild.py`, so landing independently without merge order is a real collision.

### Interlock / lineage notes

This is not a clean duplicate of the older work. #91585 (`brucexu-eth`) originally established the fail-closed direction; merged #92523 (`kshitijk4poor`, preserving Bruce's authorship) restored generic-message FTS self-heal because older SQLite needed it. #98077 supplies new forensic evidence showing why #92523's “rebuild failure will propagate, therefore canonical safety is preserved” assumption is insufficient, and this PR is a legitimate refinement of that lineage. Keep that credit chain explicit rather than treating the earlier branch as simply wrong.

Open #97843 (`SulthanZahran1`) is complementary at the turn-completion/search-degradation layer but overlaps `hermes_state.py`; it should consume the same final provenance authority after this write-boundary work settles rather than grow a second independent definition of “FTS-only.” Open #88604 (`liuhao1024`) is complementary operator/doctor classification for structural damage, not a substitute for runtime quarantine. #97940 is also the same live defect class and should be cross-linked in the closure graph rather than left as an apparently unrelated P1 after this lands.

There is a lot worth keeping here: exact-main base, small four-file footprint, green exact-head CI, explicit mixed/unscoped fail-closed behavior, and a retained old-SQLite FTS recovery control. The remaining work is at the exact boundary the incident exposed: once one participant has proof that the shared store is structurally unsafe, that proof has to survive the handle/process boundary, and the FTS exception has to be authorized by the strongest provenance SQLite actually gave us.

shivanathd:
Corroborating production evidence for the failure mode this PR closes (v0.20.6, Hermes SQLite 3.53.1, WAL, single gateway writer).

**Hermes continued accepting writes on a structurally damaged database for about ten hours.**

Timeline from the incident:

```text
05:04:46Z  last VACUUM INTO snapshot with integrity_check = ok
05:31:11Z  first "database disk image is malformed"; FTS sync disabled,
           canonical writes retried and allowed to continue
15:25Z     operator intervention (stop, forensic preserve, restore from snapshot)
```

Between those points the gateway stayed "up" in the sense that it kept serving, while every scheduled job failed with `session storage could not be written` and the transcript was not durable. That is precisely the "no obvious total persistence outage" shape described here — the degradation is invisible unless you are watching cron outcomes.

The damage was **not** FTS-only, which is the premise this PR removes. Mapping the damaged tree through `sqlite_master.rootpage`:

```text
Tree 15 page 15 cell 61..73+: 2nd reference to page ...   (dozens)
Freelist: freelist leaf count too big on page 408680
Freelist: invalid page number 167772160

select rootpage, type, name from sqlite_master;  -- 15 | table | gateway_routing
```

`gateway_routing` is a canonical table. So the generic `database disk image is malformed` was, in this case, mixed/unscoped corruption being treated as proof that only FTS was damaged — exactly the assumption this PR stops relying on.

Failing closed would have been strictly better here: ten hours of accepted-but-unusable writes produced no durable state, and the eventual recovery still had to come from an hourly snapshot taken before onset. Nothing was gained by continuing.

One adjacent observation for whoever picks this up: the first logged malformed error was already malformed — there is no preceding clean-write-then-fail — so detection time is not onset time. Bounding onset requires an external verified snapshot cadence. That matters if anyone tries to correlate these incidents against in-process events.

fangliquanflq:
Addressed the review findings:

- **Persistent quarantine (`valid_actionable`)**: structural corruption now publishes a canonical, generation-scoped authority that is observed by sibling handles, restarts, and path aliases before write callbacks can run. A successful explicit repair rotates the durable generation token and retires only quarantine state that existed when that repair began; restored quarantined generations remain blocked until repaired.
- **SQLite provenance (`valid_actionable`)**: `SQLITE_CORRUPT_VTAB` is accepted as explicit FTS provenance, contradictory known result codes override error prose, and generic/no-code corruption requires scoped integrity evidence before online FTS recovery.
- **Production timeline (`informational`)**: the reported mixed `gateway_routing`/freelist damage confirms that generic malformed-image errors cannot safely authorize FTS-only recovery; no separate code change beyond the fail-closed quarantine was needed.
- **Hosted CI repair regression (`valid_actionable`)**: generic repair fixtures without structural-quarantine markers no longer require a Hermes generation token. Matching active or retained generation quarantine still triggers token rotation and retirement.

Validation on the current tip:
- `scripts/run_tests.sh tests/state/test_fts_runtime_rebuild.py -k 'structural_quarantine_spans_handles_restarts_and_generations' -q` — 1 passed
- `scripts/run_tests.sh tests/test_state_db_repair_non_destructive.py -k 'environmental_aborts_do_not_burn_repair_ledger or stale_scratch_is_removed_before_health_check' -q` — 2 passed
- Ruff on the changed state implementation/tests — passed
- `git diff --check` — passed

The replacement GitHub Actions run is still in progress; I am not claiming hosted CI success yet.

andrexibiza (commented):
Re-reviewing exact moved head `3fa78dbfd1cfbdd450a1394facde28234c39f32d` against current `main@4f22543509d1b91dc45bcb369447126c5eb14fb7` and the new #98077 field evidence.

The two substantive P1s from my prior review are now repaired in the submitted code. The quarantine is no longer a one-handle latch: the branch publishes a durable path-wide active marker plus a retained generation marker, binds the generation in `state_meta`, rechecks the quarantine after `BEGIN IMMEDIATE`, and rotates generation only after verified offline repair. The sibling-handle/restart/refused-before-callback tests, alias-path test, retirement-failure test, and restored-old-generation test cover the important lifetime boundaries. The classifier also now composes the result-code provenance side: `SQLITE_CORRUPT_VTAB` remains FTS-scoped, contradictory known result codes do not quarantine, and generic `SQLITE_CORRUPT` needs affirmative FTS evidence before any online rebuild/fail-open. That is the right effect boundary.

There are still two landing blockers.

### P1 — the surviving seven-commit train is not green

HEAD is green: CI `33338484508`, Docker `33338484059`, and Nix `33338484067` all succeeded on `3fa78dbf...`. But acceptance belongs to every surviving commit, not only HEAD. I re-read the full train and its exact workflow receipts:

- `6736c64c...`: CI `33270260313`, Docker `33270259965`, Nix `33270259944` — green.
- `d60c5928...`: CI `33336339443`, Docker `33336339091`, Nix `33336339095` — green.
- `4f825c47...`: **CI `33336798511` failed**; Docker `33336798286` and Nix `33336798285` passed.
- `369fc574...`: **CI `33337183833` failed**; Docker `33337183382` and Nix `33337183348` passed.
- `6670f1b6...`, `fc34956f...`, and current `3fa78dbf...` subsequently have green exact-head matrices.

Those middle commits remain part of the submitted release object, so later fixes do not retroactively make the train green. Please rematerialize the final semantics as one coherent current-main commit (or otherwise rewrite the train so every surviving commit is independently green), preserving the contributor/co-author lineage. This is especially appropriate now because the branch still starts at `4209d371...` while `main` has advanced **151 commits** to `4f225435...`. Current main does touch `hermes_state.py` in a different later region, so the textual conflict surface is small, but the final state/repair object still needs current-main semantic recomposition and fresh exact-object CI.

### P1 topology — this is containment for #98077, not closure of the physical-corruption producer class

The branch is now a strong runtime containment fix for the exact dangerous behavior in #98077: generic/mixed structural corruption can no longer be laundered into FTS fail-open and canonical-write retry. It should not currently say `Closes #98077`, though.

The field-report issue explicitly separates the proven product behavior from the still-undetermined low-level producer mechanism. Since the first review, a second deployment added materially stronger evidence: corruption recurred **on SQLite 3.53.1**, with **one gateway process / one OS writer**, `synchronous=FULL`, `auto_vacuum=0`, and canonical `gateway_routing` root-page damage. That rules out "upgrade from 3.50.4" and multi-process writer concurrency as sufficient explanations. #98090 does not claim to prevent that physical corruption from being produced; it correctly prevents Hermes from continuing writes after detecting it.

Please change the topology to `Addresses #98077` (and, if appropriate, close the narrower classifier owner such as #97940 when its exact contract is satisfied) while leaving #98077 open as the forensic/root-cause record until the producer mechanism or an explicit issue split is settled. Preserve #97794 as the inverse FTS-only boundary and #86237/#98077 as physical acceptance evidence; this PR should own containment, not erase the unresolved producer question.

With the previous source blockers fixed, I do not see a new fail-open hole in the current four-file diff. The remaining work is release-object hygiene and honest topology: one current-main, every-commit-green object; containment issue linkage that does not claim the still-open corruption producer class is solved.

fangliquanflq:
I confirmed both points from the latest review:

- **Commit-train hygiene (`valid_actionable`)**: CI runs `33336798511` and `33337183833` failed on surviving intermediate commits, while the current head run is green. Producing a single current-main release object would require rewriting the published branch history. This unattended notification job is explicitly prohibited from force-pushing, so I did not rewrite, merge, commit, or push.
- **Issue topology (`valid_actionable`)**: #98077 remains open and explicitly retains the unresolved physical-corruption producer/root-cause investigation. This PR provides runtime containment, so `Addresses #98077` is accurate and `Closes #98077` is not. I did not make a partial PR-body edit while the required release-object rewrite remained blocked.

No code or PR metadata was changed in this run. The remaining blocker is a maintainer-authorized history-rematerialization path that does not violate the no-force-push constraint.

andrexibiza (commented):
Follow-up correction on exact head `3fa78dbfd1cfbdd450a1394facde28234c39f32d` against current `main@fc0a10a924ce31a7badd0d7a202dcc0779ef7942`.

The result-code and durable-quarantine direction remains right, but I missed one P1 fail-open in my prior re-review. #97940's required physical acceptance test is also still absent.

### P1 — truncated `integrity_check` output is being treated as complete FTS provenance

`SessionDB._has_fts_corruption_evidence()` runs `PRAGMA integrity_check` with no argument, collects the returned rows, then authorizes online recovery when `all(...)` returned diagnostics name Hermes FTS tables.

SQLite's own contract is that `integrity_check` reports **at most N errors and N defaults to 100**: https://www.sqlite.org/pragma.html#pragma_integrity_check

That means the returned list can be only a prefix. With mixed corruption, 100 FTS-scoped diagnostics followed by a canonical-table, index, freelist, or duplicate-page defect is indistinguishable here from FTS-only damage. The method returns `True`; `_try_runtime_fts_rebuild()` or `_enter_fts_fail_open()` can then mutate derived structures and retry the canonical write against a structurally damaged store — the exact failure class this PR must close.

Required shape: use an explicit diagnostic cap and treat a cap-hit report as **indeterminate / fail closed** (or remove generic `SQLITE_CORRUPT` online recovery entirely). A truncated report must never authorize live mutation. Add a regression at this classifier boundary where the diagnostic stream reaches the cap with only FTS-looking rows and prove no rebuild, no trigger detach, no retry, and durable structural quarantine.

### P1 acceptance — the physical covering-index counterexample is still missing

The new generic-corruption tests manufacture `sqlite3.DatabaseError`; they do not construct the access-path-dependent database required by #97940. #86237 already carries the physical fixture shape: a malformed canonical `messages` payload while a healthy covering index still answers `COUNT(*)`.

The landing object needs to drive that database through the real append boundary and prove all of the following together:

- the covering-index count still succeeds while canonical payload traversal fails;
- no FTS rebuild runs;
- no FTS trigger is removed and no fail-open marker is written;
- the failed canonical mutation is not retried;
- the discovering handle and a sibling/restarted writable handle refuse pre-effect;
- a separate physically FTS-only control still self-heals, preserving the #97794 inverse boundary.

Until that exists, mocked exception injection cannot establish the closure condition the issue now records.

### Release object and topology remain blocked

Current head CI is green ([CI 33338484508](https://github.com/NousResearch/hermes-agent/actions/runs/33338484508), [Docker 33338484059](https://github.com/NousResearch/hermes-agent/actions/runs/33338484059), [Nix 33338484067](https://github.com/NousResearch/hermes-agent/actions/runs/33338484067)), but surviving commits `4f825c47...` and `369fc574...` still have failed CI runs ([33336798511](https://github.com/NousResearch/hermes-agent/actions/runs/33336798511), [33337183833](https://github.com/NousResearch/hermes-agent/actions/runs/33337183833)). The PR is currently non-mergeable and its merge base `4209d371...` is **292 commits behind** current main. #99513 now also edits `hermes_state.py`, so the current-main successor needs an explicit merge order rather than another independent recomposition.

The honest issue topology remains `Fixes #97940` / `Addresses #98077`; #99529 is complementary reporting and shutdown/forensics work, not the classifier carrier. `Closes #98077` still overclaims the unresolved physical-corruption producer class.

fangliquanflq:
Addressed the latest review points:

- **Truncated `integrity_check` output (`valid_actionable`)**: generic corruption now runs an explicit `PRAGMA integrity_check(100)` and treats a 100-row result as indeterminate. A cap-sized FTS-looking prefix can no longer authorize an online rebuild, trigger detachment, or canonical retry.
- **Physical covering-index acceptance (`valid_actionable`)**: added a real SQLite fixture whose canonical `messages` B-tree is malformed while `COUNT(*)` still succeeds through a covering index. The write boundary now proves that no FTS rebuild runs, no trigger or stale marker changes, no mutation is retried, durable quarantine is published, and a sibling handle is refused before its callback executes. The existing physically FTS-only self-heal remains covered.
- **Current-main integration (`valid_actionable`)**: merged the latest upstream `main` normally. The only conflict was between structural-quarantine preflight and the new zero-byte startup serialization; the resolution preserves both paths, and their focused tests pass. I did not rebase or force-push.
- **Commit-train history (`valid_actionable`, blocked by delivery constraints)**: the previously failed intermediate commits still survive in published history. Making every surviving commit independently green would require rewriting that history, which this unattended flow prohibits. Current tip validation is green locally; I am not claiming the new hosted CI run has completed.
- **Issue topology (`valid_actionable`)**: the PR now uses `Fixes #97940` for the classifier defect and `Addresses #98077` for containment without claiming to solve the unresolved physical-corruption producer.

Delivered commits:
- `e11d245fc0` — fail closed on truncated integrity reports and add the physical regression.
- `6d2ccbb7ac` — merge current upstream `main` and preserve both quarantine and startup-serialization semantics.

Validation on `6d2ccbb7ac`:
- `scripts/run_tests.sh tests/state/test_fts_runtime_rebuild.py -k 'not foreign_holder_detection_proc_readlink_deleted_wal and not foreign_holder_uninspectable_process_cmdline_fallback' -q` — 27 passed
- `scripts/run_tests.sh tests/gateway/test_session.py -q` — 65 passed
- `scripts/run_tests.sh tests/test_zeroed_state_db.py -q` — 7 passed, 1 skipped
- `scripts/run_tests.sh tests/test_state_db_malformed_repair.py -q` — 21 passed, 3 skipped
- Independent scan-only review of the final conflict resolution — `perfect`
- `git diff --check hermes/main...HEAD` — passed

teknium1:
Thanks for the deep work here — the core classifier fix landed on main via PR #99652 (https://github.com/NousResearch/hermes-agent/pull/99652, merge commit f680a4dc), and closing this in favor of that.

Credit where due:
- #96038 by @diatche was submitted two days earlier with the same `SQLITE_CORRUPT_VTAB`-provenance approach for `SessionDB._is_fts_write_corruption_error`, so it was used as the salvage base with authorship preserved.
- Your PR identified the sibling site in `gateway/session.py:_is_fts_corruption_error` — that fix landed in #99652's follow-up commit, credited to this PR in the commit message. The production timeline analysis in this thread was also key evidence for the fail-closed direction.

On the durable cross-handle/restart quarantine machinery (~300 LOC): we decided not to take it. Once the classifier fails closed, writes error out and the existing repair ledger / backup / doctor path takes over; a persisted quarantine adds a risk we specifically want to avoid — a transient I/O misread (see #76280) permanently write-locking a healthy database until manual intervention. If field evidence later shows the fail-closed classifier alone isn't enough containment, that's the moment to revisit persistence, as its own scoped proposal.

Appreciate the rigor — the real-corruption fixtures and the covering-index counterexample were genuinely good test design.

Issue #38216, opened 2026-06-03
[Bug]: Hermes Desktop v40.9.3 crashes on startup on Windows 11 — 0x80000003 breakpoint exception at consistent offset


nparkison:
## Bug Description

**Hermes version:** 0.15.x (Electron 40.9.3.0)
**OS:** Windows 11 Home (build 26100)
**Hardware:** AMD Ryzen 7 9800X3D, 64GB DDR5, RX 6800 XT 16GB
**Install method:** `Hermes-Setup.exe` from hermes-agent.nousresearch.com/desktop

## Description

Hermes Desktop crashes immediately on startup every time. The app window briefly appears (Electron shell initializes, creates `%APPDATA%\Hermes` directories), then dies. Crash is 100% reproducible across multiple attempts.

## Steps to Reproduce

1. Download `Hermes-Setup.exe` from the official desktop page
2. 2. Install normally
3. 3. Launch Hermes Desktop
4. 4. App crashes within seconds
## Workarounds Attempted (none resolved)

- `hermes update --yes` before launching Desktop (per #37532 workaround) — no change
- - Deleted `~/.hermes/.gateway-planned-stop.json` (per #34597) — no change- Multiple restart attempts — identical crash every time## Windows Event Viewer Crash Logs
5 crashes logged, all identical signature:

```
Faulting application name: Hermes.exe, version: 40.9.3.0
Faulting module name: Hermes.exe, version: 40.9.3.0
Exception code: 0x80000003
Fault offset: 0x00000000076ab0b6
Faulting application path: C:\Users\<user>\AppData\Local\hermes\hermes-agent\apps\desktop\release\win-unpacked\Hermes.exe
```

Key observations:
- **Same exception code** (0x80000003 = breakpoint) across all 5 crashes
- - **Same fault offset** (0x076ab0b6) every time — deterministic crash, not random corruption
- - - **Faulting module is Hermes.exe itself**, not a dependency DLL
- - - - Crashes occurred at 8:03, 8:05 (x2), 8:45, and 8:46 AM on 2026-06-03
## Notes

- The prebuilt `Hermes-Setup.exe` still ships Electron 40.9.3.0. The fix from #37532 / PR #38158 merged into source but doesn't appear to be included in the current downloadable installer.
- - Existing Hermes CLI agent works fine on this machine (running Grok 4.3 via Discord gateway, no issues).
- - - This appears to be a bug in the prebuilt Desktop binary, not a user config issue.
[...]

PR #98092, opened 2026-08-29
fix(desktop): re-probe the sandbox once when a launch-time ACL repair succeeds

chrisluersen:
## Summary

Fixes a Windows sandbox fallback that degrades forever even after the underlying problem is repaired.

**Problem:** Hermes Desktop runs with `--no-sandbox` after repeated crash-loop boots. The only thing that ever re-probes full sandboxing is an **app version change**. But the fallback often engages because a launch-time ACL repair (granting `ALL APPLICATION PACKAGES RX` on the install dir, #38216) is **missing or failed** — and that repair happens on *every* launch. When the repair eventually succeeds (e.g. after the installer re-runs, or the user fixes permissions), the host is healthy again, yet Hermes stays pinned to `--no-sandbox` until the next app update.

**Fix:** feed the launch-time ACL-repair result into the sandbox decision. When the repair succeeds while the sticky fallback is engaged, trigger the **same one-shot re-probe** as a version change — a healed host returns to full sandboxing on the very next launch instead of waiting indefinitely.

The re-probe is **one-shot**: the fallback marker records `aclRepairProbed`, so a *still*-broken host returns straight to sticky `--no-sandbox` rather than re-crashing on every launch (same failure-tolerant behavior as the existing version-change re-probe).

## What changed

- `apps/desktop/electron/main.ts` — thread `aclRepairSucceeded` into the sandbox decision; propagate `aclRepairProbed` through the marker writes in all relaunch paths (gpu-breakpoint, renderer-crash-loop, successful-boot).
- `apps/desktop/electron/windows-sandbox-fallback.ts` — the decision state machine accepts the ACL-repair result and honors the one-shot `aclRepairProbed` marker flag.
- `apps/desktop/electron/windows-sandbox-fallback.test.ts` — 3 new unit tests covering the healed-host re-probe, the aborted re-probe returning to consumed fallback, and the failed-repair keeping today's behavior.

## Testing

`npx vitest run electron/windows-sandbox-fallback.test.ts` — **21 passed**.

## Scope

Windows-only, desktop-app only. Pure decision logic + state wiring; no change to healthy launch paths (ACL repair is skipped entirely when no fallback marker is present).

### DIFF
--- a/apps/desktop/electron/main.ts
+++ b/apps/desktop/electron/main.ts
@@ -555,10 +555,18 @@ if (PASSWORD_STORE.store) {
 // sandbox (any cause, including a manual --no-sandbox flag) — guards the
 // relaunch handlers. `windowsSandboxFallbackSticky` = the fallback machinery
 // engaged and the marker must stay `fallback` after a successful boot; a
-// manual flag alone is honored but never made sticky.
+// manual flag alone is honored but never made sticky. A successful launch-time
+// ACL repair re-probes the sandbox once (like a version change), and
+// `windowsSandboxFallbackAclProbed` keeps that one-shot consumed so a
+// still-broken host returns to sticky --no-sandbox instead of re-crashing on
+// every launch.
 let windowsSandboxFallbackActive = false
 let windowsSandboxFallbackSticky = false
 let windowsSandboxFallbackReason: SandboxFallbackReason = 'boot-loop'
+// True once the one-shot ACL-repair re-probe is in flight (or already failed)
+// for this fallback episode; the relaunch handlers carry it into the fallback
+// marker they write so a still-broken host does not re-probe every launch.
+let windowsSandboxFallbackAclProbed = false
 let windowsNoSandboxRelaunchAttempted = false
 
 if (IS_WINDOWS) {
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Heals the sticky `--no-sandbox` fallback when the underlying Windows ACL cause is repaired between launches.

**What it does**
- `apps/desktop/electron/windows-sandbox-fallback.ts:322` — when `aclRepairSucceeded && !marker.aclRepairProbed` while `state === 'fallback'`, re-probes sandbox once (same one-shot as version-change path) with `{state:'booting', reprobe:true, aclRepairProbed:true}`.
- Failure of that probe carries `aclRepairProbed:true` into fallback (`windows-sandbox-fallback.ts:349`), and `markerAfterSuccessfulBoot` preserves it (`windows-sandbox-fallback.ts:382`) so a still-broken host stays sticky and doesn't re-probe every launch.
- `apps/desktop/electron/main.ts:35` captures `aclRepairSucceeded = acl.ok` and threads it through `decideWindowsSandboxLaunch` and subsequent marker writes (`main.ts:51`, `:68`, `:96`).
- `parseSandboxMarker` and round-trip tests cover the new `aclRepairProbed: true` field strictly (only explicit `true`).

**Non-blocking notes**
- Ordering: version-change re-probe is checked first (`appVersion !== marker.version`) and does **not** consume the ACL one-shot (`windows-sandbox-fallback.test.ts:182`) — two mechanisms stay independent; correct.
- `shouldAttemptAclRepair` still gates the `icacls` call (only on prior trouble), so healthy launches skip the recursion.

No functional issues found. New tests `windows-sandbox-fallback.test.ts:107` (6 cases) pin one-shot, consumed, and sticky semantics.

chrisluersen:
Superseded by #115130 — same change, and the diff is identical (`3 files, +225/-13`). This branch was 10,539 commits behind `main`; the rebase applied cleanly, and the unit file runs **21/21 green** against current `main` @ 01382698fc versus **17 passed / 4 failed** against this branch's pre-fix source. Refreshing rather than merging as-is follows the contributing guide's warning that squash-merging from a stale branch silently reverts recent fixes.

PR #98093, opened 2026-08-29
fix(desktop): harden plugin shim against reserved-word exports and local-name collisions

chrisluersen:
## Summary

Fixes two crash classes in the desktop **plugin shim** — the ESM blob that re-exports a global namespace's live members for plugins — both triggered by minified SDK chunks.

## Bug 1: reserved word as export alias → `SyntaxError`

The shim filters export names with `/^[A-Za-z_$][\w$]*$/`, which accepts reserved words. A minified SDK chunk (proven: leaks `in` as an export alias) then produces:

```js
export const { in } = m;   // SyntaxError
```

Chromium surfaces this as a confusing `Unexpected token ')'` pointing at the chunk's line — very hard to trace back to the shim.

**Fix:** filter JavaScript reserved words from the export list.

## Bug 2: local binding collides with an export name → redeclaration error

The shim binds the namespace to a local (`const m = globalThis[...]`) then destructures exports from it. If the namespace exports a member literally named `m` (proven: the session-list-density chunk does), the generated code is:

```js
const m = globalThis.FOO;
export const { m } = m;    // Identifier 'm' has already been declared
```

**Fix:** use an unlikely internal local (`__hermes_shim_ns__`) *and* exclude it from the destructure list, so a future collision stays impossible.

## What changed

- `apps/desktop/src/sdk/runtime.ts` — add `RESERVED_WORDS` set; filter exports against it; rename the shim's local binding to `__hermes_shim_ns__` and exclude it from the destructure.

## Scope

Desktop-app only, one file. Pure string-blob generation; no behavior change for well-formed SDK chunks.

### DIFF
--- a/apps/desktop/src/sdk/runtime.ts
+++ b/apps/desktop/src/sdk/runtime.ts
@@ -25,16 +25,44 @@ export function installPluginSdk(): void {
 }
 
 /** Build a shim ESM blob that re-exports a global namespace's live members.
- *  Export names come from the namespace itself, so the list can't drift. */
+ *  Export names come from the namespace itself, so the list can't drift.
+ *
+ *  Names must be valid binding identifiers — the /^[A-Za-z_$][\w$]*$/ shape
+ *  alone is NOT enough: a minified SDK chunk can leak a reserved word as an
+ *  export alias (proven: `in`), and `export const { in } = m` is a SyntaxError
+ *  that Chromium surfaces as a confusing "Unexpected token ')'" at the chunk
+ *  line. Filter reserved words too. */
+const RESERVED_WORDS = new Set([
+  'break','case','catch','class','const','continue','debugger','default','delete',
+  'do','else','enum','export','extends','false','finally','for','function','if',
+  'import','in','instanceof','new','null','return','super','switch','this','throw',
+  'true','try','typeof','var','void','while','with','yield','let','static','await',
+  'implements','package','protected','interface','private','public'
+])
+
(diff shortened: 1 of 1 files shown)


Enough1122:

**Overall:** Hardens the plugin shim against `SyntaxError` / `Identifier has already been declared`.

**What it does**
- `apps/desktop/src/sdk/runtime.ts:17` adds `RESERVED_WORDS` set and filters `Object.keys(GLOBALS[globalKey])` by `!RESERVED_WORDS.has(name)` — prevents `export const { in } = m;` from minified chunks (proven `in` leak).
- `runtime.ts:34` uses unlikely internal binding `__hermes_shim_ns__` and excludes it from the export list (`name !== LOCAL`) so a chunk exporting `m` no longer collides (`Identifier 'm' has already been declared`).

**Non-blocking notes**
- Reserved list covers strict + future-reserved words (`let`/`static`/`await`/etc.) — good for forward-compat.
- Blob URL creation (`URL.createObjectURL`) is per install; caller revokes on navigation — no leak introduced.

No issues found. Minimal, correct.

chrisluersen:
Superseded by #110533 — same fix (the `RESERVED_WORDS` filter plus the `__hermes_shim_ns__` binding excluded from the destructure list), re-derived on current `main` with a testable `shimSource` seam and 5 vitest cases. This branch is also now conflicting against `main` (`mergeable: CONFLICTING`); ongoing work on this fix lives in #110533.

PR #98099, opened 2026-08-29
fix(skills): skill_view directory file_path + skill_manage categorized name resolution

kshitijk4poor:
## Real-world impact

Agents calling `skill_view(name, file_path=…)` on a **directory** (e.g. `references`) got a raw `[Errno 21] Is a directory: '/…/references'` OS error instead of the helpful not-found payload with the `available_files` listing. This is the exact string that recurs 28× across five months of local optimization audit logs — the background review agent hits it every time it lists a skill's support files.

Agents calling `skill_manage` with a **categorized name** (`software-development/my-skill`) got `Skill '…' not found in active profile 'default'` — even though that is the form skill_view's own ambiguity hint instructs the caller to use (`Pass the full relative path instead of the bare name (e.g., 'category/skill-name')`). Every skill_manage call that followed the hint failed, and the agent retried repeatedly — the #1 recurring error class in the audit logs (926 occurrences in one month).

## What changed

**1. `skill_view` directory request (tools/skills_tool.py)**
```python
# before
if not target_file.exists():      # a directory EXISTS → falls through
    …not-found listing…
content = target_file.read_text() # ← raises [Errno 21] on a directory

# after
if not target_file.is_file():     # directories now take the listing branch
    …not-found listing…
```
The plugin-skill sibling branch already gated on `is_file()`; this aligns the local-skill branch (same bug class, sibling call path).

**2. `skill_manage` categorized names (tools/skill_manager_tool.py)**
```python
# before — only matched the bare directory name
if skill_md.parent.name == name: …

# after — also matches the full relative path (category/skill-name)
def _matches(skill_md):
    if skill_md.parent.name == name: return True
    rel = skill_md.parent.resolve().relative_to(_skills_dir().resolve())
    return str(rel) == name
```
Gives `_find_skill`'s 12 callers (edit / patch / delete / write_file / remove_file / preflight / ledger) resolution parity with `skill_view`.

## Test plan
- [x] New regression tests fail on `main`, pass with the fix:
  - `test_view_file_path_directory_returns_available_files` (skills_tool)
  - `test_edit_existing_skill_by_categorized_path` (skill_manager_tool)
  - `test_find_skill_accepts_categorized_path` + bare-name control
- [x] Full skills test surface green: 194 passed, 1 skipped (`tests/tools/test_skill*` + `test_skills_tool.py` + background-review guard tests)
- [x] Reproduced both symptoms on a clean worktree of upstream/main before fixing

## Follow-up (review feedback)

**1ba66281a3** — addressed @kokhlo's performance note: the categorized-name match now only runs when the lookup name contains a path separator (bare-name lookups never touch the resolve machinery), and the skills root is resolved once, lazily. Also switched `str(rel)` → `rel.as_posix()` so `category/skill` lookups work on Windows, where `str(Path)` renders backslashes.

### DIFF
--- a/tests/tools/test_skill_manager_tool.py
+++ b/tests/tools/test_skill_manager_tool.py
@@ -19,6 +19,7 @@ from tools.skill_manager_tool import (
     _delete_skill,
     _write_file,
     _remove_file,
+    _find_skill,
     skill_manage,
 )
 from agent.skill_utils import (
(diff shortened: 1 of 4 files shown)


kokhlo:
Covers the two recurring errors cleanly. One performance nit on the _find_skill path:

The new categorized-name match calls skill_md.parent.resolve().relative_to(_skills_dir().resolve()) for every SKILL.md in the walk, even when the bare-name branch already matched. In a profile with 50+ skills, that is 50 resolve+relative_to calls per skill_manage invocation on a bare name.

Suggested short-circuit (zero compatibility risk, same semantics): only resolve the relative path if the bare name did not match. The existing code already works, this just skips the resolve machinery on the common case (bare-name lookups that match immediately).

kshitijk4poor:
Addressed in 1ba66281a3 — restructured so the bare directory-name check stays first and the resolve()/relative_to() machinery only runs when the lookup name actually contains a path separator; the skills root is resolved once, lazily, only when a categorized lookup happens at all. While restructuring I also noticed the original comparison used str(rel), which renders backslashes on Windows and would never match a 'category/skill' lookup there — switched to rel.as_posix(). 187 skills tests green.

PR #98101, opened 2026-08-29
fix(tui): drop the OSC 8 layer when a link label already shows the URL

### LINKED ISSUE #98091: TUI pure URL rendered as double OSC 8 hyperlink in Warp — Cmd+Click opens 2 tabs
**Environment:**
- OS: macOS 26.6.2 (arm64)
- Terminal: Warp (latest)
- Hermes: `hermes-agent` TUI (`display.interface: tui`)

**Steps to Reproduce:**
1. Launch Hermes TUI inside Warp: `hermes`
2. Have Hermes print a pure URL (e.g. `https://github.com`)
3. `Cmd+Click` the link in Warp

**Expected:** Single browser tab opens.
**Actual:** Two identical tabs open.

**Control Test:**
Outside Hermes, same Warp session:
```
echo "https://github.com"
```
`Cmd+Click` → 1 tab (correct). Inside Hermes TUI → 2 tabs.

**Root Cause (suspected):**
Hermes TUI renders a pure URL as markdown `[https://...](https://...)`, which Warp translates into two overlapping OSC 8 hyperlinks (display text layer + URL layer). Warp detects both and fires `open` twice.

**Workaround:**
Use `hermes --cli` (plain text, no OSC 8) → 1 tab.

**Suggested Fix:**
When link text equals URL, render as a single OSC 8 hyperlink (or plain URL) instead of markdown double, or emit only one `ESC]8;;URL` sequence.

liuhao1024:
## What does this PR do?

Fixes a double-open on Cmd+Click for bare URLs printed by the Hermes TUI in terminals that autodetect URLs (reported in Warp, #98091).

When Hermes TUI prints a pure URL, `ResolvedLink` renders it with a fallback label derived from the host/path (e.g. `https://github.com` shows as `github.com` until a page title resolves), and wraps that label in an OSC 8 hyperlink whose href is the same URL. Warp registers the OSC 8 hyperlink *and* its own autodetected link for the URL-shaped visible text as two overlapping click targets, so Cmd+Click opens two identical tabs. (`echo "https://github.com"` in the same Warp session opens one tab, confirming the overlay is the difference.)

The fix: when the visible label already points at the target (`normalizeExternalUrl(display) === url`), emit plain text without the OSC 8 wrapper — the terminal's own URL detection then provides a single clickable layer. Labels that differ from the target keep the OSC 8 wrapper so the exact href survives the click: authored labels, resolved page titles, and host/path labels that drop a query string (`https://example.com/1?q=1` shows as `example.com/1`).

## Related Issue

Fixes #98091

## Type of Change

## Changes Made

- `ui-tui/src/components/markdown.tsx` — in `ResolvedLink`, skip the `<Link>` (OSC 8) wrapper when the display label is URL-shaped and equivalent to the link target; extract the label element so both branches share it
- `ui-tui/src/__tests__/markdownOsc8SingleLayer.test.ts` — new regression tests pinning the single-layer behavior for bare URLs and `[url](url)` markdown, and the preserved OSC 8 href for query-dropping labels, authored labels, and resolved titles

## How to Test

1. `cd ui-tui && npx vitest run src/__tests__/markdownOsc8SingleLayer.test.ts` — 4 tests should pass (verified: `Tests 4 passed (4)`)
2. `cd ui-tui && npx vitest run src/__tests__/markdown.test.ts src/__tests__/externalLink.test.ts src/__tests__/streamingMarkdown.test.ts` — should pass (verified: 65 passed)
3. Manual (needs Warp): before the fix, the rendered output for a bare URL contains `ESC]8;id=…;https://github.com BEL github.com ESC]8;; BEL`; after the fix the same output is plain `github.com` text with no OSC 8 wrapper, so Cmd+Click opens a single tab. Observed result: the repro harness confirmed the wrapper is gone and the `?q=1` case still emits `ESC]8;id=…;https://example.com/1?q=1 BEL`
4. `npx tsc --noEmit -p tsconfig.json` — clean (exit 0); `npx eslint` on both changed files — clean

### Code

- [x] I've run `pytest tests/ -q` and all tests pass — N/A (TypeScript-only change in `ui-tui/`; ran `vitest run` instead: 1690 passed, and the 22 failures in 7 files reproduce identically on a clean `upstream/main` checkout, i.e. pre-existing environment-sensitive failures unrelated to this change)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A (behavior-preserving rendering fix, code comment documents the constraint)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-ability-guide) — N/A (OSC 8 emission is platform-independent; terminals without URL autodetection simply lose nothing since the href was redundant with the visible text)

## Screenshots / Logs

Before (repro harness, bare `https://github.com`):

```
see ESC]8;id=19nr2md;https://github.comBELgithub.comESC]8;;BEL for details
```

After: the same render emits plain `github.com` (single clickable layer via terminal autodetection), while `https://example.com/1?q=1` keeps `ESC]8;id=…;https://example.com/1?q=1BEL` so the query survives the click.

### DIFF
--- a/ui-tui/src/__tests__/markdownOsc8SingleLayer.test.ts
+++ b/ui-tui/src/__tests__/markdownOsc8SingleLayer.test.ts
@@ -1,0 +1,129 @@
+import { PassThrough } from 'stream'
+
+import { Box, renderSync } from '@hermes/ink'
+import React from 'react'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import { Md } from '../components/markdown.js'
+import { __resetLinkTitleCache } from '../lib/externalLink.js'
+import { DEFAULT_THEME } from '../theme.js'
+
+afterEach(() => {
+  __resetLinkTitleCache()
+  vi.unstubAllGlobals()
+})
+
+const ESC = String.fromCharCode(27)
+const BEL = String.fromCharCode(7)
+
+// No network in unit tests: unresolved titles fall back to the
+// hostPathLabel-derived label, which is exactly the shape whose OSC 8
+// wrapping stacks on the terminal's own URL autodetection (#98091).
+const stubUnresolvedTitles = () => {
+  vi.stubGlobal('fetch', vi.fn().mockRejectedValue(new Error('offline')))
+}
+
+// The renderer emits `ESC ] 8 ; id=<group> ; url BEL` (the id param groups
+// wrapped lines), so match the url with an optional id segment.
+const osc8For = (url: string) =>
+  new RegExp(`${ESC}\\]8;(?:id=[^;${BEL}]*;)?${url.replace(/[.*+?^${}()|[\]\\]/g, String.raw`\$&`)}${BEL}`)
+
+const renderPlain = (text: string, width = 120) => {
+  const stdout = new PassThrough()
+  const stdin = new PassThrough()
+  const stderr = new PassThrough()
+  let output = ''
+
+  Object.assign(stdout, { columns: 80, isTTY: false, rows: 24 })
+  Object.assign(stdin, { isTTY: false })
+  Object.assign(stderr, { isTTY: false })
+  stdout.on('data', chunk => {
+    output += chunk.toString()
+  })
+
+  const instance = renderSync(
+    React.createElement(Box, { width }, React.createElement(Md, { t: DEFAULT_THEME, text })),
+    {
+      patchConsole: false,
+      stderr: stderr as NodeJS.WriteStream,
+      stdin: stdin as NodeJS.ReadStream,
+      stdout: stdout as NodeJS.WriteStream
+    }
+  )
+
+  instance.unmount()
+
+  return output
+}
+
+describe('URL-shaped labels emit a single clickable layer (#98091)', () => {
+  it('does not wrap a bare URL whose label equals the target in OSC 8', () => {
+    stubUnresolvedTitles()
+
+    const output = renderPlain('see https://github.com for details')
+
+    // The fallback label (hostPathLabel) is `github.com`, which the
+    // terminal autodetects as a link by itself; an OSC 8 wrapper on top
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Fixes double-tab open on Cmd+Click for bare URLs in Warp (and similar terminals with URL autodetection).

**What it does**
- `ui-tui/src/components/markdown.tsx:164` — `ResolvedLink` now emits plain `<Text>` (no OSC 8 `<Link>`) when `normalizeExternalUrl(display) === url` — i.e. the fallback label is URL-shaped (`github.com` for `https://github.com`). Warp was stacking its autodetected link on top of our OSC 8 hyperlink.
- Keeps OSC 8 when label drops query string (`example.com/1` for `https://example.com/1?q=1`) or when authored label / fetched title differs (`ui-tui/src/components/markdown.tsx:164` predicate), so exact href survives.
- Tests `ui-tui/src/__tests__/markdownOsc8SingleLayer.test.ts:65` cover bare URL, markdown `[url](url)`, query-dropping, and authored/title cases via `renderSync` + `ESC ]8 url BEL` regex.

**Non-blocking notes**
- `normalizeExternalUrl` must be symmetric with `display` derivation (hostPathLabel) — the test's stubbed `fetch` rejection correctly exercises the fallback path.
- No change to `Link` component itself — minimal blast radius.

No functional issues found.

liuhao1024:
Thanks for the review. Confirmed the predicate stays symmetric with the display-label derivation — the bare-URL and `[url](url)` cases drop the OSC 8 layer while query-dropping and authored/title labels keep it, which is exactly what the four `markdownOsc8SingleLayer` cases pin. No changes to `Link` itself keeps the blast radius minimal.

PR #98102, opened 2026-08-29
fix(cron): refuse cross-vendor model/provider pairs at create/update time

Kyzcreig:
## Problem

A cron job can be pinned to a `model` and a `provider` from **different vendors** — e.g. `model='gpt-5-codex'` + `provider='anthropic'`. `cronjob(action='create')` accepts it, `create_job()` persists it, and then every single fire returns an opaque HTTP 400 ("the model field names a different vendor model than this endpoint serves").

This is not a transient outage. Every retry and every future tick fails identically, so the job is **dead on arrival** — and because the failure lives in per-run delivery errors rather than at the call site, it's typically only noticed after the job has already burned a string of scheduled runs. Catching the pair at create/update time turns a silent recurring cron failure into one immediately actionable error.

## Fix

Refuse a provably cross-vendor pair at admission time, next to the existing `_validate_cron_base_url` guard in `tools/cronjob_tools.py`:

- **`model_provider_vendor_mismatch(model, provider)`** — infers each side's vendor from a conservative name-prefix map (`_MODEL_VENDOR_PREFIXES` / `_PROVIDER_VENDOR_PREFIXES`) and returns `(model_vendor, provider_vendor)` only when both are known and differ.
- **Fail-open by design.** An unrecognized model or provider name yields vendor `None` and is **not** flagged. This rule exists to catch obvious mis-pairs, not to maintain a model catalog — a brand-new model name simply falls through and is allowed. That's the property that keeps this from becoming a maintenance burden as vendors ship names.
- **CREATE:** checked before `create_job_with_scheduler_registration()` persists anything, so a rejected pair leaves no job behind.
- **UPDATE:** checked against the **effective post-update pair**, reusing the same merged `eff_provider` the `base_url` re-validation already computes. An update may supply only `model` or only `provider`, so repointing an openai-pinned job at an anthropic provider is only detectable after merging the update over the stored job.

The error names both inferred vendors and states explicitly that retrying will not help:

```
model 'gpt-5-codex' (openai) cannot run on provider 'anthropic' (anthropic) — this pair
fails every run with HTTP 400 (the model field names a different vendor model than this
endpoint serves; retrying will not help). Pick a provider that serves this model's vendor,
or clear one side of the pin.
```

## Tests

New `TestModelProviderVendorConsistency` in `tests/tools/test_cronjob_tools.py`, following the file's existing `monkeypatch.setattr("cron.jobs.CRON_DIR", ...)` fixture convention:

- predicate: mismatch detected; matching pair not flagged; **fail-open** on unrecognized model, unrecognized provider, and `None`/`None`
- create rejected on a cross-vendor pair, error names both vendors, and `cronjob(action='list')` confirms **nothing was persisted**
- create allowed on a matching pair, and allowed when either side is unrecognized
- update rejected when **only the provider** changes, and when **only the model** changes
- update allowed when the pair stays consistent, and for unrelated fields (rename)

**Verification on this base:**
- `pytest tests/tools/test_cronjob_tools.py tests/cron` → **1106 passed, 10 skipped** on the branch.
- **RED-proved:** with the two hook call sites removed (helpers left in place), exactly the 3 rejection tests fail (`create_rejected_on_cross_vendor_pair`, `update_rejected_when_only_provider_changes`, `update_rejected_when_only_model_changes`); restoring them returns green. The tests fail for the reason they claim.

## Notes

- Ports a fix from our fork ([ANG-Ventures/hermes-agent#641](https://github.com/ANG-Ventures/hermes-agent/pull/641)), re-authored against this tree rather than cherry-picked. Adaptation: the fork's create path builds a preview-job dict for a shared admission-gate helper this tree doesn't have, so the checks are hooked directly into `cronjob()`'s create/update branches beside the existing `base_url` validation. The fork also handles a `{"model": ..., "provider": ...}` dict shape for `model`; here `model` is normalized to a plain string by `_normalize_optional_job_value`, so only the string shape is handled.
- No new config knobs, no new env vars, no new tools — two guard calls plus a pure predicate.

### DIFF
--- a/tests/tools/test_cronjob_tools.py
+++ b/tests/tools/test_cronjob_tools.py
@@ -7,6 +7,7 @@ from tools.cronjob_tools import (
     _scan_cron_prompt,
     check_cronjob_requirements,
     cronjob,
+    model_provider_vendor_mismatch,
 )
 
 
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Admission-time guard against dead-on-arrival cron jobs with cross-vendor `(model, provider)` pins.

**What it does**
- `tools/cronjob_tools.py:153` adds conservative prefix maps `_MODEL_VENDOR_PREFIXES` / `_PROVIDER_VENDOR_PREFIXES` and `_vendor_of` + `model_provider_vendor_mismatch` (`tools/cronjob_tools.py:189`).
- `_validate_model_provider_vendors` (`tools/cronjob_tools.py:202`) returns an actionable error (`"fails every run with HTTP 400 ... retrying will not help"`) when both sides' vendors are provably different; unknown names fail-open (`None`).
- `cronjob(action="create")` validates the normalized pair before persistence (`tools/cronjob_tools.py:234`); `cronjob(action="update")` validates the effective post-update pair (`tools/cronjob_tools.py:253`), catching `model`-only or `provider`-only updates that would otherwise repoint an `openai`-pinned job at `anthropic`.
- Tests `tests/tools/test_cronjob_tools.py:733` cover mismatch detection, fail-open on unknowns, create rejection/persistence-zero, and update of effective pair.

**Non-blocking notes**
- Prefix coverage is deliberately narrow (`codex → openai`, `o1/o3/o4 → openai`) — brand-new model families fall through as `None` and are allowed; correct fail-open trade-off.
- Error message includes both model/Provider strings and their inferred vendors — actionable.
- No catalog to maintain; when a vendor adds a new naming pattern, add a prefix entry (e.g. future `sonnet` still `claude*` covered).

No functional issues found.

Kyzcreig:
Your non-blocking note turned out to be blocking. Verified F01 independently and it's a genuine regression this PR introduced, not an inherited one — fixed in `31e40409`.

> the new name-prefix vendor guard rejects supported raw custom-provider names on create/update, even when runtime resolution matches the accepted `custom:`-qualified spelling exactly

Confirmed at the previous head `9d46a915`. Resolving the same model under raw `provider=openai-compatible` and qualified `provider=custom:openai-compatible` produces **identical** runtime identity — both resolve to `provider=custom`, same endpoint, same model, same transport. The admission guard then read the raw label's first segment, mistook `openai-compatible` for vendor `openai`, and rejected `(meta, openai)` as cross-vendor — while admitting the `custom:` spelling. Switching an existing job from qualified to raw also rejected before any write. The base branch accepts both spellings, so I introduced this.

The root defect was comparing a *label* to a *vendor*. The fix resolves the provider identity first (reusing runtime's named-custom lookup and builtin alias resolver, without fetching credentials) and only then applies the vendor check, keeping legacy vendor-label coverage for names with no registered identity.

Verification, driving the real admission predicate and temp-file persistence for both spellings:

```
RED    (unfixed head)      3 failed, 2 passed
GREEN  new class           5 passed   (both spellings admitted identically; qualified→raw update; 
                                       a genuine builtin cross-vendor pair STILL rejected)
GREEN  existing admission  11 passed
GREEN  test_cronjob_tools  87 passed
REVERT control             3 failed, 2 passed  (original guard restored → new tests go red)
```

The "still rejected" case is there deliberately so the fix can't over-correct into admitting real cross-vendor pairs — which is the thing the guard exists for.

Thanks for flagging it as judgment-call rather than dismissing it; it was the right instinct. The PR is `CONFLICTING` against current main, which is mine to rebase separately.

Issue #48974, opened 2026-06-19
[project-local] #4 Skills: scan-time injection quarantine


alt-glitch:
## Parent
Part of EPIC #48970. Depends on #1. Can parallel #2/#3.

## Goal
Scan-time injection quarantine for committed project skills — the ONLY defense once auto-load is sticky.

## Why
Committed/pulled `.hermes/skills/<name>/SKILL.md` is read straight from disk by the index + slash scanners (`agent/prompt_builder.py:1335-1352`, `agent/skill_commands.py:369-410`). `skills_guard` only runs in the hub install path, so a repo checkout bypasses it entirely. Regression test class proving the threat: `tests/cron/test_cron_prompt_injection_skill.py`.

## Scope
- Scan project SKILL.md content for high-confidence prompt-injection directives before it enters the index / slash map.
- Quarantine (don't load) on high-confidence hit; surface findings in the consent prompt (#3) so the user sees what they're approving.
- Re-scan when the content hash changes (ties into the #3 fingerprint).

## Acceptance
- A project skill carrying an injection payload is quarantined and NOT loaded; the consent prompt shows the finding. Mirror the `test_cron_prompt_injection_skill` regression shape.

## Standard
Security hardening (`type/security`). E2E with a real malicious-skill fixture, not mocked.

teknium1:
Shipped in PR #88643 (merged 6e22d265835). Every project skill now runs through the same `skills_guard` scanner as Skills Hub installs before entering the index — content-hash cached under `~/.hermes/cache/project_skill_scans/` (never inside the repo), re-scanned automatically when content changes.

A `dangerous` verdict quarantines the skill: excluded from the system-prompt index, `skills_list`, and slash commands via a single iteration chokepoint (`iter_project_skill_files`), and `skill_view` refuses it by name with an explanatory error. Scanner failure fails closed. Verified E2E against a real injection fixture (live verdict `dangerous`, 6 findings: prompt_injection_ignore, deception_hide, invisible_unicode, hermes_env_access, read_secrets_file, dump_all_env) — mirroring the `test_cron_prompt_injection_skill` regression shape as specified here.

Issue #43282, opened 2026-06-10
In-process LRU cache key in build_skills_system_prompt ignores SKILL.md content changes (mtime/size)


willingning-coder:
## Bug Description

`build_skills_system_prompt()` uses an in-process LRU cache (`_SKILLS_PROMPT_CACHE`) whose key does **not** include any fingerprint derived from SKILL.md file content. When a skill file is modified externally (manual edit, another session, sync tool), the cache continues serving the stale `<available_skills>` block — including stale descriptions, tags, and metadata — until the gateway process restarts or `clear_skills_system_prompt_cache()` is explicitly called.

The disk snapshot layer (Layer 2) already validates via mtime/size manifest, but Layer 1 short-circuits before Layer 2 is ever consulted.

## Affected Code

`agent/prompt_builder.py` — `build_skills_system_prompt()`, lines ~1109-1131:

```python
# Layer 1: in-process LRU cache
cache_key = (
    str(skills_dir.resolve()),
    tuple(str(d) for d in external_dirs),
    tuple(sorted(str(t) for t in (available_tools or set()))),
    tuple(sorted(str(ts) for ts in (available_toolsets or set()))),
    _platform_hint,
    tuple(sorted(disabled)),
)
with _SKILLS_PROMPT_CACHE_LOCK:
    cached = _SKILLS_PROMPT_CACHE.get(cache_key)
    if cached is not None:
        _SKILLS_PROMPT_CACHE.move_to_end(cache_key)
        return cached
```

No file-content-derived value (e.g. manifest hash) is included in the key.

## Why This Matters

`skill_manage` correctly calls `clear_skills_system_prompt_cache(clear_snapshot=True)` after modifying a skill, so agent-initiated changes are fine. But **external modifications** — manual file edits, another session in a parallel surface (Telegram, WebUI, Discord), git pull, sync tools — leave the in-process cache stale.

This is distinct from:
- **#14535** — that issue is about `disabled_skills` config not being in the key (config state)
- **#41303** — that issue is about `/reload-skills` not detecting modified content (command behavior)

This issue is specifically about the **cache key** itself lacking a content-fingerprint component.

## Reproduction

1. Start a gateway session (loads skills into `_SKILLS_PROMPT_CACHE`)
2. In a separate terminal, modify a skill description on disk:
   ```bash
   sed -i "s/description: old/description: NEW/" ~/.hermes/skills/research/my-skill/SKILL.md
   ```
[...]

PR #98103, opened 2026-08-29
fix(skills): key the project-skill quarantine cache on content, not path

irakli1980h-blip:
## Summary

`_PROJECT_QUARANTINE_CACHE` (`agent/skill_utils.py`) memoised the project-skill scan verdict **per directory path**. `scan_skill_cached` underneath it already re-scans whenever the bundle hash changes, but the memo above it never did — so the first verdict a process computed was the only one it could ever see.

That breaks in both directions:

**1. A repaired skill stays quarantined.** The scanner flags prose — a bare `mkfs` mention, a command-substitution example — you rewrite the file, a fresh scan on disk returns `safe`, and `skill_view` keeps answering *"quarantined: it will not load until the repo's skill content changes and passes a re-scan"*. The content **did** change and it **did** pass; only the in-process memo disagrees, until restart. The advice in the error message is unreachable.

**2. The same stale memo caches an _allow_.** A skill that was clean when first seen and then edited to something dangerous inside the same process keeps loading. The gate from #48974 silently stops applying to it. This is the direction that matters for the feature's purpose.

## Fix

Include `full_content_hash(skill_dir)` in the cache key, so the memo answers for the exact bundle it was computed from. If hashing raises, skip the cache entirely rather than reuse an unrelated verdict — the fail-closed behaviour on scanner failure is unchanged.

The hash covers the whole bundle, so poisoning or cleaning a `references/` file re-evaluates the verdict too, not just edits to `SKILL.md`.

## Test plan

Two additions to `TestQuarantine`, both **failing before this change**:

- `test_repair_clears_quarantine_without_cache_clear` — direction 1. It deliberately does **not** call `_project_quarantine_cache_clear()`. The existing `test_rescan_after_content_change` does call that hook, which is exactly why it passed throughout and the bug went unnoticed.
- `test_regression_reintroduces_quarantine_in_same_process` — direction 2, so the content key cannot regress into a one-way unlock.

Verified on `main` at `4209d371aa`:

```
# without the fix, tests present
2 failed
  test_repair_clears_quarantine_without_cache_clear      AssertionError: assert True is False
  test_regression_reintroduces_quarantine_in_same_process AssertionError: assert False is True

# with the fix
tests/agent/test_project_skills.py .......................... 24 passed
tests/tools/test_skills_guard.py ........................... 36 passed
+ tests/agent/test_skill_utils.py, tests/agent/test_skill_commands.py
                                                            112 passed
```

Full `tests/agent/` was run both ways: **171 failures on stock `main`, 171 with this patch** (the run with the patch shows 172/5503 vs 171/5502 — the delta is this PR's own new test passing, plus one pre-existing flake in `test_title_generator` that is unrelated to skills, passes 5/5 in isolation and 3/3 as a file, and shares no code path with this change). Set-differencing the two failure lists yields **no new failures introduced**.

## Notes

Found while writing a project-local skill that documents security-scanner false positives — the skill described the trap and then fell into it, and the documented repair loop did not work. Related but distinct: #43282 is the same class of bug in a different cache (`_SKILLS_PROMPT_CACHE` in `prompt_builder.py`, key lacks a content fingerprint); this one is `_PROJECT_QUARANTINE_CACHE` and carries the security-gate consequence above.

Worth a separate look: `/reload-skills` (`agent.skill_commands.reload_skills`) routes through `iter_project_skill_files` and therefore the same chokepoint, but never invalidates this cache either — so even the command a user reaches for after fixing a skill could not clear a stale quarantine before this change.

### DIFF
--- a/agent/skill_utils.py
+++ b/agent/skill_utils.py
@@ -834,9 +834,11 @@ def get_scan_ordered_skills_dirs() -> List[Path]:
 # write artifacts into the user's checkout).
 
 _PROJECT_SCAN_SOURCE = "project-local"
-# (skill_dir_resolved) -> quarantined bool, keyed per-process; scan_skill_cached
-# already re-scans on content change via the bundle hash, this only avoids
-# re-reading the attestation JSON on every index/list/view call in one run.
+# (skill_dir_resolved, bundle_hash) -> quarantined bool, keyed per-process;
+# scan_skill_cached already re-scans on content change via the bundle hash,
+# this only avoids re-reading the attestation JSON on every index/list/view
+# call in one run. The hash belongs in the key so that repairing a quarantined
+# skill takes effect immediately instead of at the next process start.
 _PROJECT_QUARANTINE_CACHE: Dict[str, bool] = {}
 
 
@@ -855,12 +857,25 @@ def is_quarantined_project_skill(skill_md) -> bool:
     """
     skill_dir = Path(skill_md).parent
     try:
-        key = str(skill_dir.resolve())
+        path_key = str(skill_dir.resolve())
     except OSError:
-        key = str(skill_dir)
-    cached = _PROJECT_QUARANTINE_CACHE.get(key)
-    if cached is not None:
-        return cached
+        path_key = str(skill_dir)
+    # Key on content, not just path: an author who FIXES a quarantined skill
+    # mid-session must be able to load it again. scan_skill_cached already
+    # re-scans on content change (bundle hash), but a path-only key pinned the
+    # stale "dangerous" verdict for the life of the process, so the repaired
+    # skill stayed invisible until restart.
+    try:
+        from tools.skills_guard import full_content_hash
+
+        key = f"{path_key}\0{full_content_hash(skill_dir)}"
+    except Exception:
+        # Hashing failed — do not reuse any cached verdict for this dir.
+        key = None
+    if key is not None:
+        cached = _PROJECT_QUARANTINE_CACHE.get(key)
+        if cached is not None:
+            return cached
     try:
         from tools.skills_guard import scan_skill_cached
 
@@ -883,7 +898,8 @@ def is_quarantined_project_skill(skill_md) -> bool:
             exc_info=True,
         )
         quarantined = True
-    _PROJECT_QUARANTINE_CACHE[key] = quarantined
+    if key is not None:
+        _PROJECT_QUARANTINE_CACHE[key] = quarantined
     return quarantined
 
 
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Fixes stale-quarantine bug where a repaired project skill stayed blocked for the life of the process. Small, well-tested change.

**What it does**
- `agent/skill_utils.py` keys `_PROJECT_QUARANTINE_CACHE` on `f"{path}\0{hash}"` instead of path alone. `full_content_hash(skill_dir)` is the bundle hash already used by `scan_skill_cached`, so editing a skill mid-session now produces a new cache key and re-evaluates.
- On hash failure (`Exception`) sets `key=None` and skips cache read/write — avoids pinning a stale verdict when hashing is unavailable; re-scans every call until hashing recovers.
- Updates type hint comment and stores per-content verdict only when `key is not None`.
- Tests add both directions: `test_repair_clears_quarantine_without_cache_clear` (evil → benign becomes visible without `_project_quarantine_cache_clear()`) and `test_regression_reintroduces_quarantine_in_same_process` (benign → evil re-blocks), plus `iter_project_skill_files` visibility checks.

**Non-blocking notes**
- `full_content_hash` is imported inside `is_quarantined_project_skill` — fine for lazy import, but consider hoisting for clarity; failure path intentionally bypasses cache which could add scan cost if hashing is persistently broken — acceptable since that path is exceptional.
- Cache is unbounded per-content (`path\0hash`). In normal use collisions are bounded by number of skill versions in one process; consider `lru` or size cap if long-lived gateway edits many skills in place, but current scope is small.
- `from tools.skills_guard import full_content_hash` inside try — if the import itself fails, `key=None` means no caching at all; good fail-open for scanning but slightly more I/O — intentional trade-off.

No blocking issues.

supplefrog:
One cache-identity edge appears to remain. `full_content_hash()` uses the existing bundle digest, whose directory stream is formed from each relative path plus a NUL plus its bytes, without length-framing each record or separately identifying record/file type. The scanner's classification is path- and extension-sensitive, so that digest is a compatibility/content identifier but not necessarily an unambiguous identity for the exact scan input. Keying the outer verdict cache on it can therefore retain a stale verdict for a different scan-relevant file layout.

Would it make sense to use a separate, versioned internal scan-cache fingerprint with explicit record framing (and keep the public bundle/content hash unchanged for hub compatibility), then key both scanner and quarantine verdict caches on that identity? Focused tests could assert that filename/extension and record-boundary changes always change the internal fingerprint.

This is source inspection against head `25433622eed450b97f895ebcd8647f67e1cfa957`. I am treating the scanner as a heuristic quarantine mechanism, not claiming it is a complete security boundary.

Issue #98028, opened 2026-08-29
Deliberate client absence (app closed / PC asleep) kills the in-flight turn: 20s ws_orphan_reap grace ends overnight remote work


Navlem:
## Summary

On a remote (VPS) gateway, **deliberately leaving the client — closing Hermes Desktop or sleeping the PC — now kills the in-flight turn** with a bare `Operation interrupted.` This is a behavior change, not a transport bug: before #14b50f5edd a mid-turn session survived client absence indefinitely, which made "start a long task, close the laptop, read it in the morning" a working pattern. It no longer is, and the 20 s default grace means there is no window in which an overnight absence can survive.

This is the *intentional-absence* case. It is distinct from the existing reports, which are all about **unwanted** disconnects on a client that is still present:

- #94196 / #69940 — WS drops every few minutes while the user is actively using the app (fix = stop dropping).
- #95327 (closed) — Windows backend respawn kills the turn; asks for a better placeholder + fewer respawns.
- #84207 — the placeholder gives no feedback (UX of the message).
- #85016 — same *shape* as this, but scoped to `api_server` SSE, not the desktop WS gateway.

Fixing every one of those still leaves this broken: a user who *intends* to disconnect for hours loses the work anyway.

## Environment

- Hermes Agent v0.20.6 (2026.8.27), git install, upstream `3f36c87e`
- Linux VPS (6.8.0-137-generic), Python 3.11.15, profile `default`
- Desktop app on a separate machine → `hermes dashboard --host 0.0.0.0` on the VPS
- `dashboard.ws_orphan_reap_grace_s`: default `20.0`; `ws_ping_interval` / `ws_ping_timeout`: `20.0` / `20.0`

## Root cause

`14b50f5edd` ("fix(tui-gateway): interrupt turns after websocket disconnect", 2026-08-19, first released in `v2026.8.27`) inverted the mid-turn guard in the orphan reaper.

Before, `_ws_session_is_orphaned` refused to reap a running session:

```python
if session.get("running"):
    return False          # mid-turn → never reaped, parked until the client returns
return session.get("transport") is _detached_ws_transport
```

After, `_schedule_ws_orphan_reap._reap` interrupts it instead (`tui_gateway/server.py` ~1421):

```python
elif current.get("running"):
    ...
    if not current.get("_client_gone_interrupt_requested"):
        current["_client_gone_interrupt_requested"] = True
        interrupt_session = current          # → _interrupt_session_turn(...)
```

[...]

PR #98106, opened 2026-08-29
fix(gateway): fence orphan timers and reconnect interrupt races

BearHuddleston:
## Summary

Re-scoped on current `main` (`1e69c12b64dba4090eddffeae27f5a147068d34b`) after #100504 merged the activity-based fix for healthy detached turns. This PR now contains only the remaining orphan-timer and reconnect/interrupt race fixes.

- Fence dispatched callbacks and their continuations by timer identity, so an obsolete detachment cannot remove or replace a newer detachment's timer.
- Register the first orphan timer while holding the detachment claim, closing the disconnect/reconnect/redetach scheduling window.
- Serialize lazy/unpersisted resume, activation, and prompt transport rebind with the orphan interrupt claim; reject a settling interrupt with the existing `4009` response and revalidate live-session identity.
- Avoid recursive acquisition of the non-reentrant resume lock in the eager-resume double-check.
- Cancel a concurrent cold-resume winner's timer only after guarded reuse accepts reattachment; a rejected reconnect must not strand interrupt-settlement polling.

## Policy and scope

The previous revision's fixed 12-hour running-turn allowance and `ws_orphan_running_grace_s` setting are **removed**. This revision does not change configuration or documentation defaults. It preserves #100504's policy: fresh activity protects detached running turns, stale/missing activity is eligible for interruption, and `ws_orphan_activity_stale_s: 0` retains interrupt-at-grace behavior. Existing delegation protection, bounded interrupt settlement/force-reap, explicit Stop, and sidecar `close_on_disconnect` behavior remain intact.

Follow-up to #100504; related to #98028. No merging or unrelated PR changes.

## Verification

All Python test runs used the canonical `scripts/run_tests.sh` runner with the development interpreter, clean environment, temporary Hermes home, and per-file subprocess isolation.

- New deterministic regression file: **19 passed** on this head. Applied to unchanged upstream: **15 failed, 4 passed**, reproducing the stale callback/continuation/initial-registration races and reconnect/eager-resume failures. Tests were written and observed failing before their fixes.
- Independent review identified a concurrent cold-resume cancellation regression. Added the `cold_resume_claim` regression and observed it fail, fixed the cancellation point, then repeated the independent review: **passed**, no security concerns or logic errors.
- Focused orphan/reconnect/claim coverage: **45 passed, 0 failed**.
- Broad affected suite below: **1,702 passed, 1 failed** across 115 files. The sole failure, `tests/test_tui_gateway_server.py::test_model_options_preserves_canonical_custom_row_after_agent_init`, also fails in a separate unchanged-upstream worktree (**1,683 passed, 1 failed**, 114 files). No new broad-suite failures. This is not a claim that the full repository suite passed.
- Ruff on all six changed files, Python bytecode compilation, `scripts/check_compat_pointers.py`, and `git diff --check`: **passed**.

```bash
scripts/run_tests.sh tests/tui_gateway/test_ws_orphan_races.py -q
scripts/run_tests.sh tests/tui_gateway/test_ws_orphan_races.py tests/test_tui_gateway_server.py \
  -k 'orphan or reconnect_cannot or claim_or_reuse_live' -q
scripts/run_tests.sh tests/tui_gateway/ tests/test_tui_gateway_server.py \
  tests/test_tui_gateway_server_crash_history.py tests/test_tui_gateway_ws.py \
  tests/test_tui_gateway_event_replay.py tests/test_tui_gateway_loop_noise.py \
  tests/test_tui_gateway_queue_on_busy.py tests/test_ws_keepalive_config.py \
  tests/test_pty_keepalive_ws.py -j 8 -q
```

Local verification is reported above; GitHub CI status is tracked by the checks on this PR.

### DIFF
--- a/tests/test_tui_gateway_server.py
+++ b/tests/test_tui_gateway_server.py
@@ -5439,7 +5439,7 @@ def test_resume_rebind_cancels_pending_ws_orphan_reap(monkeypatch):
 
 
 def test_claim_or_reuse_live_winner_cancels_pending_reap(monkeypatch):
-    """A resume that reuses the live winner cancels the winner's pending reap."""
+    """The winner's pending reap is cancelled only once guarded reuse is accepted."""
     cancelled = []
 
     class _Timer:
@@ -5472,6 +5472,17 @@ def test_claim_or_reuse_live_winner_cancels_pending_reap(monkeypatch):
         )
 
         assert live == ("winner-sid", winner)
+        assert "winner-sid" in server._pending_ws_reaps
+        assert cancelled == []
+        assert winner["transport"] is server._detached_ws_transport
+
+        transport = object()
+        monkeypatch.setattr(server, "current_transport", lambda: transport)
+        ctx = server._Resume(1, {"omit_messages": True}, "stored-claim")
+        response = server._resume_reuse_live(ctx, *live)
+
+        assert response["result"]["session_id"] == "winner-sid"
+        assert winner["transport"] is transport
         assert "winner-sid" not in server._pending_ws_reaps
         assert len(cancelled) == 1
     finally:
--- a/tests/tui_gateway/test_ws_orphan_races.py
+++ b/tests/tui_gateway/test_ws_orphan_races.py
@@ -1,0 +1,159 @@
+"""Orphan callbacks own only their detachment, never a later reconnect."""
+
+from contextlib import nullcontext
+import threading
+from types import SimpleNamespace
+from unittest.mock import Mock
+
+import pytest
+
+from tui_gateway import server
+
+
+@pytest.mark.parametrize("phase", ["before_callback", "before_continuation", "before_initial_timer", "cold_resume_claim"])
+def test_obsolete_orphan_cannot_replace_new_detachment(monkeypatch, phase):
+    timers = []
+
+    class Timer:
+        def __init__(self, delay, callback):
+            self.callback = callback
+            timers.append(self)
+
+        def start(self):
+            pass
+
+        def cancel(self):
+            pass  # A dispatched callback can still execute after cancel().
+
+    sid = "generation-race"
+    session = dict(transport=server._detached_ws_transport, running=True,
+                   agent=SimpleNamespace(get_activity_summary=lambda: {"seconds_since_activity": 0}))
+    monkeypatch.setattr(server, "_sessions", {sid: session})
+    monkeypatch.setattr(server, "_pending_ws_reaps", {})
+    monkeypatch.setattr(server.threading, "Timer", Timer)
+    monkeypatch.setattr(server, "_WS_ORPHAN_REAP_GRACE_S", 20)
+    monkeypatch.setattr(server, "_WS_ORPHAN_ACTIVITY_STALE_S", 600)
+    monkeypatch.setattr(server, "_session_has_active_delegations", lambda *a: False)
+    if phase == "before_initial_timer":
+        transport = object()
+        session["transport"] = transport
+        newest = None
+
+        class DisconnectLock:
+            def __enter__(self):
+                pass
+
+            def __exit__(self, *args):
+                # A new client reconnects and drops as the old disconnect
+                # releases its claim, before any out-of-lock scheduling.
+                nonlocal newest
+                server._cancel_ws_orphan_reap(sid)
+                server._schedule_ws_orphan_reap(sid)
+                newest = timers[-1]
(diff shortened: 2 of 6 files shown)


Enough1122:

**Overall:** Preserves detached browser/Desktop turns through disconnects while adding a bounded 12h safety net for wedged turns. Correctly hardens the WS-orphan reaper against stale-timer races. Largest diff in this chunk (38KB) but well-tested.

**What it does**
- `hermes_cli/config_defaults.py` adds `dashboard.ws_orphan_running_grace_s` (43200s default) and separates idle reap (`ws_orphan_reap_grace_s: 20s`) from running-turn grace. `0` disables both (`_resolve_ws_orphan_running_grace` returns 0).
- `tui_gateway/server.py` rewrites `_schedule_ws_orphan_reap` to snapshot `ws_orphan_running_grace_s` per-detachment, store `_client_gone_running_deadline` (monotonic, `None` when unlimited) and `_client_gone_reap_token` (opaque `object()` identity). `_reap` now: drains only its own generation (`_pending_ws_reaps.get(sid) is not timer` + token check), defers running turns until deadline (`min(reap_grace, remaining)`), then runs the existing bounded interrupt-then-poll chain. Timers co-exist via identity checks, not `cancel()` alone (which cannot stop an already-dispatched callback).
- `_cancel_ws_orphan_reap` clears both deadline and token; `_close_sessions_for_transport` and `_live_session_payload` cancellation paths also clear them.
- `_close_sessions_for_transport` no longer immediately interrupts a running detached turn — it schedules the reap and lets the turn finish within its deadline.
- `tui_gateway/methods_session.py:session.activate` now holds `_session_resume_lock` while checking `_client_gone_interrupt_requested` and returns `4009 "session disconnect interrupt settling"` if the reaper has already claimed the interrupt — prevents rebinding a transport mid-settlement.
- Docs/example config updated to reflect idle vs running semantics.

**Non-blocking notes**
- Running-grace timer delay is `min(_WS_ORPHAN_REAP_GRACE_S, remaining)` on each poll — ensures the bound is a real maximum, not just noticed at the next idle tick. Good.
- `with _sessions_lock:` around deadline computation then re-check under lock avoids TOCTOU for racing reconnect/re-detach. Stale-continuation test (`test_ws_orphan_reap_stale_continuation_cannot_replace_new_deadline`) covers interleaved reconnect correctly.
- `_resolve_ws_orphan_running_grace` reads `load_config()` on each detachment outside locks — keeps lifecycle locks uncontended; second deadline check after load avoids re-attaching stale state if a reconnect raced the load.
- Complexity is high (token + deadline + poll chain). The five new `test_ws_orphan*` tests (stale callback, removed-before-reap, bounded interrupt) materially reduce regression risk — keep them.

No blocking issues; monitor in production for orphaned `running` sessions that outlive the 12h bound and for `4009` error rate.

AlejandroAkbal:
Thank you for working on this. The current disconnect behavior is one of the biggest practical limitations for me: I regularly connect from an iPhone to an always-on remote Hermes backend, and an ordinary network or app transition shouldn't stop a long-running turn. Keeping the work alive and letting clients reconnect to the same state would make Hermes dramatically more useful day to day. This looks like a strong step in that direction, and I'd really love to see it land.

kshitijk4poor:
Thanks @BearHuddleston — this landed via #104137, which cherry-picks your commit as-is (authorship preserved; `1ac9019` on that branch) onto current `main` and adds four small follow-ups on top:

- the identical 4007/4009 guard you added at four sites → one `_reattach_refusal` helper in `session_lifecycle.py`
- `session.activate` now builds its payload *after* releasing `_session_resume_lock` (the Ink TUI calls activate without `omit_messages`, so the DB history read was happening under the process-wide lock); the rebind itself stays under the lock as you had it
- the lazy/unpersisted resume path and `_live_session_payload` share one `_rebind_live_transport`
- that helper lives in `session_lifecycle.py` next to the reap machinery rather than in the `server.py` facade

Your race analysis and tests are unchanged — all 19 cases are in the salvage and 16 of them still go red against `main`'s source. Auto-merge is armed on #104137; closing this one in its favour.

@Enough1122 — the 12h `ws_orphan_running_grace_s` your review discusses was already dropped from this PR's later revision (superseded by #100504's activity-based policy); nothing in the salvage changes config or defaults.
@AlejandroAkbal — the "keep the turn alive while the phone reconnects" behaviour is #100504 (already on `main`); this PR closes the reconnect races around it.

PR #95433, opened 2026-08-26
fix(compression): retry a stalled summary on the fallback chain (#78981)

shauneccles:
Fixes the "compression not falling back after an aborted commit" leg of #78981.

## Problem

A stalled context-compression summary (connection open, no tokens) never raises, so the auxiliary client's fallback handling — which only runs from its exception path — is unreachable. The host's progress-aware timeout aborts the worker and degrades to "continue without compression", never attempting `auxiliary.compression.fallback_chain`.

## Fix

When `run_compress_context_with_progress_timeout` aborts a stalled worker, it now re-runs the summary once, pinned to the first `auxiliary.compression.fallback_chain` entry, before degrading:

- `resolve_compression_fallback_route()` resolves the first structurally complete chain entry into explicit `call_llm` route arguments (provider/model/base_url/api_key/api_mode/timeout, key via the existing `_fallback_entry_api_key`).
- The pin is a single-use `ContextVar` in `agent/context_compressor.py`, consumed by the summary call. Context is copied per pool worker, so it reaches the retry's call chain and cannot leak into the detached stalled worker or the compressor's own main-model retry.
- The retry runs on a fresh fence minted through a host factory (`_publish_new_fence`), so a `/stop` during the retry still serializes against the live commit boundary. It is skipped entirely if a hard interrupt is already pending.
- `on_timeout` (and its cooldown bookkeeping) fires only after the fallback attempt also fails, so it cannot suppress the retry it precedes.

`agent/auxiliary_client.py` is intentionally untouched — no overlap with #80122 / #81274, which fix the false-progress signal.

## Behaviour

- No `fallback_chain` configured → unchanged: abort and continue without compression.
- Fallback succeeds → compression recovered on the fallback route.
- Fallback also stalls or errors → degrade exactly as before, with a WARNING logged.

## Tests

`tests/agent/test_compression_stall_fallback_78981.py` — 11 tests, RED-verified before GREEN:

- stalled summary attempts the configured fallback chain
- fallback-that-also-stalls degrades after exactly one attempt
- route pin injection / single-use semantics
- no-chain and hard-interrupt-skip paths

Ran 5x consecutively (timing-sensitive file). Surrounding compression suites green (46 files, 449 tests). `ruff` clean.

## Scope / known limitations

- Only the first chain entry is pinned — one bounded extra attempt is the stall budget. If that entry *errors* (not stalls), the aux client's own chain walk takes over from there.
[...]

PR #98108, opened 2026-08-29
compression: skip a stalled primary summary route after fallback-recovered stalls (#95879)

### LINKED ISSUE #95879: compression: escalate a stalled primary summary route, not just raising failures
## Problem

PR #95433 retries a stalled context-compression summary once on the first `auxiliary.compression.fallback_chain` entry before degrading. That fixes the stall itself — a silent stall never raises, so the auxiliary client's own fallback handling couldn't see it — but it leaves one gap.

When the fallback recovers successfully, the stall path returns before `on_timeout`, so `record_timeout_failure` never runs. The per-session escalation ladder therefore never learns that the primary route stalled. A permanently half-dead summarizer — one that holds the connection open and emits nothing — is retried at full speed every compaction: each turn burns one full `context_timeout_seconds` idle window on the dead primary before the fallback saves it, forever.

## Why the obvious fix is wrong

Recording the stall in the existing ledger would disable *all* compression, not just the primary. `record_timeout_failure` → `_record_compression_failure_cooldown` sets `_summary_failure_cooldown_until`, which makes `_generate_summary` return `None` immediately and makes `compress()` refuse to run at all. A successful fallback recovery would then suppress compression entirely for the cooldown window — including the fallback route that just worked. That's strictly worse than the idle window it was meant to save, and it re-breaks the very ordering #95433 relies on (fallback runs before `on_timeout` precisely so the cooldown can't suppress the retry).

## Proposed fix

A separate, route-specific stall ledger rather than a change to the shared cooldown ladder:

- count consecutive primary-route stalls that were recovered by the fallback,
- after N such recoveries, skip the primary and make the first `fallback_chain` entry the first attempt for a cooldown window,
- re-probe the primary on some schedule (or on the next successful compression),
- reset the counter when a primary attempt actually produces tokens.

Per-session and persisted like the existing cooldown state, so a gateway restart doesn't reset it mid-escalation.

## Context

- PR: #95433 — "fix(compression): retry a stalled summary on the fallback chain"
- Parent bug: #78981 (silent provider stall)
- Config: `auxiliary.compression.fallback_chain`, `context_timeout_seconds`
- Code: `run_compress_context_with_progress_timeout` / `_retry_compression_on_fallback_chain` (agent/conversation_compression.py), `record_timeout_failure` / `_record_compression_failure_cooldown` (agent/context_compressor.py), `_on_timeout` (run_agent.py)

## Cost of not doing this

Bounded but real: one idle window (~120s default) of dead latency per compaction for any session whose primary summary route has gone half-dead. Compression still succeeds via the fallback, so nothing is dropped — only the latency is paid, every time.

Shotflame:
## Summary

Fixes #95879. A permanently half-dead primary summarizer (connection open, zero tokens) was retried at full speed every compaction: the fallback-RECOVERED path in `run_compress_context_with_progress_timeout` returns before `on_timeout` fires, so the shared cooldown ladder never learned the primary stalled, and every compaction burned a full idle window (default 120s) on the dead primary forever.

## Mechanism

- `on_primary_stall` callback fired only on the fallback-recovered path (before `on_timeout`), so the shared route-agnostic ladder stays untouched (recording a recovery there would suppress the healthy fallback and re-break #95433).
- Route-specific persisted ledger on `ContextCompressor` (`_primary_stall_streak`, `_primary_stall_skip_until`) via session model_config keys: no schema migration, restart-safe.
- At `compression.primary_stall_skip_threshold` (default 2) fallback-recovered stalls, a bounded skip-primary window (`compression.primary_stall_skip_seconds`, default 600s) arms; the host pre-pins `fallback_chain[0]` as the FIRST attempt instead of probing the half-dead primary.
- Window lapse re-probes the primary; a stall re-arms a fresh window (bounded cadence); an unpinned (real primary) summary success clears the ledger durably.
- Telemetry: `failure_class=primary_route_stall_recovered` + `primary_stall_streak` make the stall visible in the compression telemetry stream.

## Tests

15 new tests in `tests/agent/test_compression_primary_stall_escalation.py` (hook fires on recovered stall / never on healthy route, ledger increment to gate, pre-pin makes fallback the first attempt, pre-pinned fallback stall degrades once, telemetry, below-threshold no skip, window expiry re-probe + re-arm, unpinned success resets, pinned success keeps streak, persistence round-trip, session-rotation carry, durable reset). Full regression suite passes (one pre-existing timing flake in `test_compression_review_76354` fails identically on base).

Reviewed and APPROVED in kanban review (t_4d407de1): no blocking comments; 6 non-blocking notes recorded.

### DIFF
--- a/agent/context_compressor.py
+++ b/agent/context_compressor.py
@@ -106,6 +106,38 @@ _PINNED_ROUTE_FIELDS: tuple[str, ...] = (
     "timeout",
 )
 
+# #95879: persisted keys for the route-specific primary-stall ledger. The
+# ledger is deliberately SEPARATE from the shared summary-failure cooldown
+# ladder (record_timeout_failure / _summary_failure_cooldown_until): that
+# ladder is route-agnostic, so recording a fallback-recovered stall there
+# would suppress ALL compression — including the fallback that just worked —
+# and re-break the #95433 ordering (fallback runs before on_timeout precisely
+# so the cooldown can't suppress the retry). Values ride the session's
+# model_config JSON (same merge discipline as proactive-prune rearm), so no
+# schema migration is needed and a restart cannot disarm mid-escalation.
+_PRIMARY_STALL_STREAK_KEY = "primary_stall_streak"
+_PRIMARY_STALL_SKIP_UNTIL_KEY = "primary_stall_skip_until"
+_PRIMARY_STALL_DEFAULT_SKIP_THRESHOLD = 2
+_PRIMARY_STALL_DEFAULT_SKIP_SECONDS = 600.0
+
+
+def _compression_config_value(key: str, default: Any):
+    """Read one ``compression.*`` config value, falling back to ``default``.
+
+    Mirrors ``resolve_context_compression_timeouts``'s read of the same
+    section: fresh read per call (the events are rare), never raises.
+    """
+    try:
+        from hermes_cli.config import load_config
+
+        raw = load_config()
+        cfg = raw.get("compression", {}) if isinstance(raw, dict) else {}
+        if isinstance(cfg, dict) and key in cfg and cfg[key] is not None:
+            return cfg[key]
+    except Exception:
+        pass
+    return default
+
 
 @contextlib.contextmanager
 def pin_summary_route(route: Optional[Dict[str, Any]]):
@@ -2303,6 +2335,8 @@ class ContextCompressor(ContextEngine):
         self._structural_no_op_backoff_until = 0.0
         self._prellm_skip_count = 0
         self._fallback_compression_streak = 0
+        self._primary_stall_streak = 0
+        self._primary_stall_skip_until = 0.0
         self._verify_compaction_cleared_threshold = False
         self._last_compression_made_progress = False
         self._summary_failure_cooldown_until = 0.0  # transient errors must not block a fresh session
@@ -2607,6 +2641,8 @@ class ContextCompressor(ContextEngine):
         self._structural_no_op_backoff_until = 0.0
         self._prellm_skip_count = 0
         self._fallback_compression_streak = 0
+        self._primary_stall_streak = 0
+        self._primary_stall_skip_until = 0.0
         self._verify_compaction_cleared_threshold = False
         self._last_compression_made_progress = False
         self._summary_failure_cooldown_until = 0.0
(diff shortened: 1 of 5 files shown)


Enough1122:

**Overall:** Fixes permanent half-dead primary summarizer stall (#95879). A dedicated, persisted route-specific ledger counts fallback-recovered stalls and pre-pins fallback as first attempt after threshold, saving one full idle window (default 120s) per compaction forever. Carefully separates this from the shared cooldown ladder to preserve #95433 ordering.

**What it does**
- `agent/context_compressor.py` adds `_PRIMARY_STALL_STREAK_KEY` / `_PRIMARY_STALL_SKIP_UNTIL_KEY` (persisted via `model_config` JSON, same channel as proactive-prune rearm, no schema migration). New methods: `_load/_persist_primary_stall_state`, `record_primary_route_stall_recovered` (increments streak, arms `skip_until = now+600s` when `streak >= threshold(=2)` and window lapsed), `record_primary_route_success` (clears on unpinned success only — respects `_SUMMARY_ROUTE_CONSUMED` echo so a main-model retry after pinned fallback error does not clear), `_primary_stall_gate_tripped`/`should_skip_primary_route` (refreshes durable state before deciding, so sibling-process clear is honored).
- `agent/conversation_compression.py:run_compress_context_with_progress_timeout` adds `on_primary_stall` callback fired only on fallback-recovered path before return; `on_timeout` (shared ladder) deliberately NOT fired there. Also emits `primary_stall_streak` in telemetry when `failure_class=="primary_route_stall_recovered"`.
- `run_agent.py` pre-checks `should_skip_primary_route()` and `resolve_compression_fallback_route()`, pre-pins first fallback chain entry via `pin_summary_route` as the *first* attempt; sets `stall_fallback=False` for that attempt so it degrades once via shared ladder rather than looping on same route.
- `hermes_cli/config_defaults.py` adds `compression.primary_stall_skip_threshold` (2) and `primary_stall_skip_seconds` (600s).
- Session-rotation carry in `on_session_start` / `bind_session_state` preserves ledger across compression child rotation and restart (wall-clock epoch ↔ monotonic conversion).
- Exhaustive test file `tests/agent/test_compression_primary_stall_escalation.py` (15+ cases): hook firing, streak→gate, pre-pin, healthy-route unchanged, skip-window expiry reprobe, pinned vs unpinned success, persistence/restart, rebind-to-other-session isolation, rotation carry.

**Non-blocking notes**
- `_load_primary_stall_state` stores epoch in `model_config` but in-memory is monotonic (`skip_until = monotonic + (epoch - time.time())`). Correct conversion, but clock skew between wall and monotonic could shift window by skew delta — negligible for 600s window.
- `record_primary_route_success` guards with `_SUMMARY_ROUTE_CONSUMED.get()` — verify that context-var is set on the same task/thread as `_generate_summary`; otherwise a retry could incorrectly clear ledger. Tests cover pinned success keeps ledger.
- Config reads per stall/success are rare events — `_compression_config_value` fresh `load_config()` each time is fine.
- Consider logging `should_skip_primary_route` decision at DEBUG when gate is tripped but fallback chain is empty (pre_pin_route would be None, falls back to primary anyway).

Well-isolated performance/hang fix with strong persistence semantics.

PR #98109, opened 2026-08-29
fix(doctor): skip browser-use warning when backend is off

SharpieCoder1776:
## What does this PR do?

When `browser.backend` is explicitly `off`, Hermes uses the built-in agent-browser `browser_*` tools. `hermes doctor` still listed the unused `browser-use` toolset as `⚠ browser-use (system dependency not met)` because that toolset has no env vars and its check_fn returns false.

This treats an intentional opt-out as a missing dependency. Doctor already has the same pattern for kanban/honcho in `_apply_doctor_tool_availability_overrides`.

## Related Issue

No existing issue. Reproduced on a seat with `browser.backend: 'off'` and a working agent-browser install.

## Type of Change

## Changes Made

- `hermes_cli/doctor.py`: skip the `browser-use` unavailable row when `get_browser_backend() == BACKEND_DISABLED`
- `tests/hermes_cli/test_doctor.py`: suppress when backend is off; keep unavailable when backend is `browser-use`

## How to Test

1. Set `browser.backend: off` in `~/.hermes/config.yaml`
2. Run `hermes doctor`
3. Confirm Tool Availability has no `⚠ browser-use (system dependency not met)` while `browser` / `browser-cdp` stay healthy
4. `uv run --extra dev pytest tests/hermes_cli/test_doctor.py::TestDoctorToolAvailabilityOverrides -v`

Red-green: stashing the doctor.py change makes `test_suppresses_browser_use_when_backend_is_off` fail (browser-use remains in unavailable).

### Code

- [ ] I've run `pytest tests/ -q` and all tests pass (ran the targeted doctor override class, 4 passed; full suite not run)
- [x] I've tested on my platform: macOS 27.0 Darwin arm64 (M5 Max)

- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (uses existing `get_browser_backend()`)

### DIFF
--- a/hermes_cli/doctor.py
+++ b/hermes_cli/doctor.py
@@ -364,6 +364,16 @@ def _apply_doctor_tool_availability_overrides(available: list[str], unavailable:
             if "honcho" not in updated_available:
                 updated_available.append("honcho")
             continue
+        # browser-use is intentionally disabled when browser.backend is "off"
+        # (seat uses agent-browser browser_* tools instead). Suppress noise.
+        if name == "browser-use":
+            try:
+                from tools.browser_use_cli import BACKEND_DISABLED, get_browser_backend
+
+                if get_browser_backend() == BACKEND_DISABLED:
+                    continue
+            except Exception:
+                pass
         updated_unavailable.append(item)
     return updated_available, updated_unavailable
 
--- a/tests/hermes_cli/test_doctor.py
+++ b/tests/hermes_cli/test_doctor.py
@@ -226,6 +226,50 @@ class TestDoctorToolAvailabilityOverrides:
         assert available == []
         assert unavailable == [kanban_entry]
 
+    def test_suppresses_browser_use_when_backend_is_off(self, monkeypatch):
+        """backend: off means agent-browser is the stack — not a missing dep."""
+        monkeypatch.setattr(doctor, "_honcho_is_configured_for_doctor", lambda: False)
+        import tools.browser_use_cli as bu
+
+        monkeypatch.setattr(bu, "get_browser_backend", lambda: bu.BACKEND_DISABLED)
+        browser_use_entry = {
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Suppresses false-positive `⚠ browser-use (system dependency not met)` warning in `hermes doctor` when `browser.backend: off` intentionally opts into agent-browser. Minimal, follows existing kanban/honcho pattern.

**What it does**
- `hermes_cli/doctor.py:_apply_doctor_tool_availability_overrides` adds `if name=="browser-use"` branch: calls `get_browser_backend()` and `continue`s (drops from `unavailable`) when it equals `BACKEND_DISABLED`. Wrapped in `try/except` so import or config errors degrade to showing the warning rather than hiding it.
- `tests/hermes_cli/test_doctor.py` adds two cases: backend `off` removes `browser-use` from unavailable while leaving other entries (discord) untouched; backend `browser-use` keeps it unavailable. Tests mock `_honcho_is_configured_for_doctor` to isolate.

**Non-blocking notes**
- `from tools.browser_use_cli import BACKEND_DISABLED, get_browser_backend` inside the loop — each unavailable entry triggers an import lookup. `browser-use` is typically one entry, so cost is negligible; consider hoisting import to module top with lazy fallback if `doctor` is imported frequently.
- Broad `except Exception: pass` hides `get_browser_backend` failures — intentional to avoid masking real dependency issues, but could hide misconfiguration that *should* warn. Current choice (show warning on error) is safe.
- `get_browser_backend()` may read config/env; ensure it has no side effects when called from `doctor` — assuming pure getter per `tools/browser_use_cli.py`.

No blocking issues.

PR #98111, opened 2026-08-29
fix(agent): Z.AI overload backoff was pinned to one model on one path, so every other rail gave up in 15s

rodrigogs:
The adaptive long-backoff tier for Z.AI overload 429s (30/60/90/120s) only engaged when the base URL contained `api.z.ai/api/coding/paas/v4` **and** the model name contained `glm-5.2`. Every other Z.AI rail gets the same 429 and got none of the policy.

## What that cost, measured on a live board

Workers on `https://api.z.ai/api/paas/v4/` running `glm-4.7-flash` take the identical response:

```
HTTP 429: The service may be temporarily overloaded, please try again later
{'code': '1305', 'message': 'The service may be temporarily overloaded, please try again later'}
```

Because the predicate did not recognise them, `max_retries` stayed at the default 3, the retry loop never reached the long tier, and the run gave up after three short waits — 2.7s, 5.2s, 5.5s: roughly **15 seconds of patience against a transient service window**. Nine runs across three cards ended that way in one evening.

`zai_coding_overload_retry_ceiling`'s own docstring describes the schedule going dead when the ceiling equals `short_attempts`. For these rails the ceiling was never raised at all, so it was dead for a second reason.

## What changes

`is_zai_coding_overload_error` now keys on what actually identifies the condition: **a 429 from a Z.AI host carrying code 1305 / "temporarily overloaded"**. The model pin and the coding-plan path pin are gone. `model` stays in the signature for call-site symmetry and future per-model policy.

**The narrowness that matters is kept.** Z.AI's other 429s must still fail fast so the fallback chain moves on: `Weekly/Monthly Limit Exhausted` (a spent plan window does not improve in 120 seconds) and code `1302` `Rate limit reached for requests` both stay outside the policy, and so does every non-Z.AI provider.

**Failover behaviour is untouched.** The detection only raises the retry ceiling; `_should_fallback` already routes an `overloaded` classification to the next chain entry after two attempts, so this changes nothing while another rail is available. It matters exactly where it should — on the last entry, when the chain is exhausted and waiting is the only move left.

**The host is parsed, not substring-matched.** `"z.ai" in base_url` would also accept `api.z.ai.example.com` and hand a stranger's endpoint a two-minute wait. Bare `host/path` values still resolve, since call sites are not guaranteed to pass a scheme.

The names keep their `coding` (the behaviour was first seen on the Coding Plan); renaming would touch call sites for no behavioural gain, and the docstrings now say the shape belongs to the service.

## Verification

`tests/test_retry_utils.py` goes from 11 to **21** passing; `tests/agent/test_error_classifier.py` is unchanged and still fully green (**142** on the current rebase against `main`).

The predicate had **no tests** — the file carries a gap where they would sit, right after the `_zai_overload_error` fixture, which is how a gate this narrow survived. The gap is filled: the historical coding-plan shape as a regression, the standard-endpoint shape as the new behaviour, and the refusals as the substance — quota wall, code 1302, another provider's identical text, a lookalike host, a missing base URL, and a non-429 status.

Mutation-verified: **6 mutants, 0 survivors** — restoring the model pin, restoring the coding-path pin, dropping the overload-text check, dropping the host check, replacing the parsed host with a naive substring, and dropping the status check. Each is killed by a different assertion.

Related: #97623 fixes the other half of what these nine runs produced — a throttled worker exiting 0 and being recorded as a protocol violation.


Enough1122:

**Overall:** Widens Z.AI overload long-backoff from a single `api.z.ai/api/coding/paas/v4 + glm-5.2` shape to any `*.z.ai` host + any model with 429/1305 overload signature. Fixes live-measured premature give-up after 3 short retries (~15s) on `api.z.ai/api/paas/v4` with `glm-4.7-flash`, unlocking the 30/60/90/120s schedule.

**What it does**
- `agent/retry_utils.py` adds `_is_zai_host(base_url)` using `urlsplit(...).hostname` (lowercased) with bare-host fallback (`raw.split("/",1)[0].split(":",1)[0]`) — prevents `z.ai in base_url` lookalike matches (`api.z.ai.example.com`, `notz.ai`, `z.ai.attacker.test`). Checks `host == "z.ai"` or `endswith(".z.ai")`.
- `is_zai_coding_overload_error` now: early `status_code != 429` and `not _is_zai_host`, then narrow payload check (`"1305" in text or "temporarily overloaded" in text`) only — removes `base` and `model` pinning. `model` param retained for call-site symmetry. Retains narrow shape so quota 1113 and rate-limit 1302 429s still fail fast (needed for chain failover).
- `adaptive_rate_limit_backoff` doc updated to reflect generic Z.AI overloaded shape.
- Tests cover: historical shape regression, new multi-model/path match, quota/rate-limit negatives, lookalike host negatives, bare-host matching, missing base, non-429 overload text negative, and end-to-end `long_waits == [30,60,90,120]` on standard endpoint.

**Non-blocking notes**
- `urlsplit("api.z.ai/api/paas/v4").hostname` is `None` (no scheme), so fallback path `raw.split("/",1)[0]` → `api.z.ai` → host check works — verified for scheme-less call sites. For `https://api.z.ai:443/v1`, `hostname` strips port correctly via stdlib.
- Broad `or "temporarily overloaded" in text` could match a non-Z.AI provider that reuses that phrase, but host gate restricts to `*.z.ai` so blast radius is limited to Z.AI service.
- Name `is_zai_coding_overload_error` is now historical (comment says so); consider alias `is_zai_overload_error` to avoid future confusion, but rename is breaking — keeping name is pragmatic.

Well-scoped reliability fix with strong negative tests.

rodrigogs:
Rebased onto current `main` (no conflicts; `agent/retry_utils.py` and `tests/test_retry_utils.py` had not diverged). `tests/test_retry_utils.py`: 21 passed; `tests/agent/test_error_classifier.py`: 142 passed; ruff clean.

On the alias note (`is_zai_overload_error`): kept the existing name on purpose. The predicate is re-exported through `agent/conversation_loop.py`'s lazy-import table and imported by `agent/turn_recovery.py`; adding a second public name for the same function would leave two spellings for reviewers and callers to keep in sync, with no behavioural gain. The docstring records that `coding` is historical, which is the only place the ambiguity can bite.

rodrigogs:
**Rebased onto `upstream/main` (`5dea46d13d`); GitHub reported this PR `CONFLICTING`/`dirty` before the push.**

**What the force-push overwrites:** the previous remote head `2c337b43a` is replaced by `48f330e10`. Same single commit, same 2 files (`agent/retry_utils.py`, `tests/agent/test_retry_utils.py`).

Worth recording, because the two did not agree: this branch **replayed cleanly**, one commit, no conflict — and a merge of the same head into `upstream/main` (`git merge-tree --write-tree`) also reported none. GitHub nonetheless answered `mergeable: false` / `mergeable_state: dirty` on three reads spread over a minute before the push. I did not resolve anything, because there was nothing to resolve; the rebase is the only intervention, and it settles the report — GitHub now answers `mergeable: true` on the new head.

The rebase touched nothing in the diff: `_is_zai_host()` (host shape — `host == "z.ai"` or `endswith(".z.ai")`, with the bare-host fallback), the widened `status_code != 429` early-out and the narrow `1305` / `temporarily overloaded` payload check in `is_zai_coding_overload_error()`.

**Verified on the rebased head** (`scripts/run_tests.sh`): `tests/agent/test_retry_utils.py` → **21 passed, 0 failed**.

PR #98112, opened 2026-08-29
fix(kanban): count request_review/request_changes as terminal worker exits

### LINKED ISSUE #98107: [Bug]: the worker exit guard does not count kanban_request_review as a terminal action - it nudges the worker into a kanban_complete that cannot succeed
## Summary

A worker that correctly ends its turn with `kanban_request_review` is told it
"tried to exit without kanban_complete/kanban_block" and is nudged to finish. The
only tool the nudge accepts is `kanban_complete`, which on a card already in
`review` fails with "unknown id or already terminal". The worker burns turns
arguing with a guard that is wrong.

## How this differs from the existing exit-guard issues

This is not the iteration-budget case (#23216), the text-response finish (#27178),
the rc=0 protocol_violation revert (#47893) or the Windows early-exit (#42289).
Here the worker *did* call a lifecycle tool, the transition *did* succeed, and the
card *is* in a correct state. The guard simply does not have `request_review` in
its list of acceptable terminal actions.

## Impact

Cosmetic in the best case, expensive in the worst: on a local backend each extra
turn is a full prompt of tool schemas and a full decode. It also trains the wrong
lesson - a model that learns to satisfy the guard will call `kanban_complete`
instead of `request_review` and skip review entirely.

## Observed

From a real worker log (profile `dev`, card in `review` after a successful
`kanban_request_review`):

    👀 preparing kanban_request_review…
    ⚠️ Kanban worker tried to exit without kanban_complete/kanban_block - nudging to finish
    ✔ preparing kanban_complete…
    ⚡ kanban_co  0.0s [could not complete t_4a177ae4 (unknown id or already terminal)]
    ✔ preparing kanban_complete…
    ⚡ kanban_co  0.0s [could not complete t_4a177ae4 (unknown id or already terminal)]

### Expected Behavior

worker is able to ends its turn with `kanban_request_review`

### Actual Behavior

The model's own reply to the nudge states the situation correctly:

    "In the Hermes Kanban protocol, a task in the review column is considered a
     terminal state for the current worker (the implementation phase is finished).
     Any attempt to call kanban_complete on a task that is already in review or
     done results in the error: could not complete ... (unknown id or already
     terminal)."

It was nudged twice and made two impossible calls before exiting.

### Affected Component

Other

N/A (CLI only)

### Debug Report

```shell
## Environment

Hermes Agent v0.20.5, Linux aarch64, local OpenAI-compatible backends,
`kanban.review_dispatch: true`.
```

### Operating System

Ubuntu 24.04 (DGX Spark)

### Python Version

_No response_

### Hermes Version

_No response_

```shell

```

_No response_

## Suggested fix

Add `kanban_request_review` (and, for review workers, `kanban_request_changes`) to
the set of actions that satisfy the worker exit contract, in the guard and in the
kanban-worker skill text that states the contract. Both are documented terminal
transitions for the run that performs them: `request_review` ends the
implementation run, `request_changes` ends the review run.

liuhao1024:
## What does this PR do?

The kanban worker stop guard only accepted `kanban_complete` / `kanban_block` as run-ending tools. But the worker system prompt (KANBAN_GUIDANCE §5) teaches `kanban_request_review(summary=..., metadata=...)` as the normal ender when a card needs review, and `kanban_request_changes(reason=...)` as the reviewer's ender for a review run. A worker that followed the prompt and ended its turn right after a successful `kanban_request_review` was nudged ("tried to exit without kanban_complete/kanban_block"), and the only tool the nudge taught — `kanban_complete` — fails on an already-review card with "unknown id or already terminal", so the worker burned turns on impossible calls.

Both transitions are documented run-enders in the board state machine, so the guard and the dispatcher already disagree about them today:

- `kanban_request_review` moves the card out of `running` into the review lane, so the dispatcher's clean-exit check (which only flags tasks still `running`) never treats it as a protocol violation — only the in-loop guard does.
- `kanban_request_changes` ("Finish an active review run and route the task back for rework") closes the review run; review workers spawn through the same `_default_spawn` path and run under `HERMES_KANBAN_TASK`, so the guard fires for them too.

The fix adds both tools to `_TERMINAL_KANBAN_TOOLS` in `agent/kanban_stop.py` (the single source the conversation-loop guard consults) and updates the synthetic nudge to teach every legal run-ender instead of only complete/block, with an explicit "if a lifecycle call already succeeded, do not repeat it — pick the one legal transition for the card's current column" line to break the impossible-`kanban_complete` loop from the issue report.

## Related Issue

Fixes #98107

## Type of Change

## Changes Made

- `agent/kanban_stop.py` — added `kanban_request_review` and `kanban_request_changes` to `_TERMINAL_KANBAN_TOOLS` (with a comment explaining why both are run-enders); updated the module docstring; rewrote nudge step 2 to list all four legal run-enders and to warn against repeating an already-successful lifecycle call.
- `agent/conversation_loop.py` — guard comment block and the `_emit_status` message now say "a terminal kanban tool" instead of enumerating only complete/block (comment + status-string only, no logic change).
- `tests/agent/test_kanban_stop.py` — added `test_no_nudge_after_kanban_request_review` (the issue's exact scenario), `test_no_nudge_after_kanban_request_changes` (review-worker ender), and `test_nudge_lists_all_run_enders` (pins that the nudge teaches all four tools so the taught-tool set can't silently drift from the guard set again).

## How to Test

1. `pytest tests/agent/test_kanban_stop.py -q` — Observed result: `6 passed` (3 pre-existing + 3 new).
2. Adjacent kanban suites: `pytest tests/agent/test_kanban_stop.py tests/hermes_cli/test_kanban_block_kinds.py tests/hermes_cli/test_kanban_core_functionality.py -q` — Observed result: `31 passed, 1 skipped` (the skip is pre-existing).
3. Manual shape of the reported scenario, now covered by `test_no_nudge_after_kanban_request_review`: a message history containing a successful `kanban_request_review` tool call returns `None` from `build_kanban_stop_nudge`, so the worker's turn ends normally instead of being nudged toward an impossible `kanban_complete`; before the fix, the same history produced a nudge (`kanban_request_review` was not in the terminal set).
4. `ruff check` on the three changed files passes; `ruff format --check` is clean for the new code (pre-existing main-only reformat diffs in untouched regions of `conversation_loop.py` / `test_kanban_stop.py` were left alone deliberately).

### Code

- [x] I've run `pytest tests/ -q` and all tests pass — ran the targeted kanban suites (`tests/agent/test_kanban_stop.py`, `tests/hermes_cli/test_kanban_block_kinds.py`, `tests/hermes_cli/test_kanban_core_functionality.py`: 31 passed, 1 skipped); this change touches only the guard's tool-name set, nudge string, and one status string, with no other call sites
- [x] I've tested on my platform: macOS 15 (darwin 25.4.0, arm64); change is platform-independent (pure Python policy set)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — module docstring in `agent/kanban_stop.py` and the guard comment updated to state the full terminal contract
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — N/A (no config keys changed)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — N/A (no platform-specific code; new tests use monkeypatched env vars only)
- [x] I've updated tool descriptions/schemas if I changed tool behavior — N/A (tool behavior unchanged; only the stop guard's acceptance set)

### DIFF
--- a/agent/conversation_loop.py
+++ b/agent/conversation_loop.py
@@ -8407,7 +8407,9 @@ def run_conversation(
                     continue
 
                 # ── Kanban worker terminal-tool stop guard ─────────────
-                # Workers must end with kanban_complete / kanban_block.
+                # Workers must end their run with a terminal board tool
+                # (kanban_complete / kanban_block / kanban_request_review /
+                # kanban_request_changes — see agent/kanban_stop.py).
                 # Models sometimes narrate the next step ("Let me write the
                 # report") and stop with finish_reason=stop — a clean exit
                 # that the dispatcher records as protocol_violation. Nudge
@@ -8443,7 +8445,7 @@ def run_conversation(
                     )
                     agent._emit_status(
                         "⚠️ Kanban worker tried to exit without "
-                        "kanban_complete/kanban_block — nudging to finish"
+                        "a terminal kanban tool — nudging to finish"
                     )
                     # Same finalizer contract as verify-on-stop: clear
                     # final_response while continuing so a later budget
--- a/agent/kanban_stop.py
+++ b/agent/kanban_stop.py
@@ -1,7 +1,10 @@
 """Turn-end guard for kanban workers.
 
-Kanban workers must end with ``kanban_complete`` or ``kanban_block``. Models
-(especially GLM / Qwen families) sometimes narrate the next step
+Kanban workers must end their run with a terminal board tool:
+``kanban_complete``, ``kanban_block``, ``kanban_request_review`` (ends the
+implementation run; the card moves to the review lane), or
+``kanban_request_changes`` (ends a review run; the card returns for rework).
+Models (especially GLM / Qwen families) sometimes narrate the next step
 ("Let me write the report now") and stop with ``finish_reason=stop`` and no
 tool calls. Hermes treats that as a clean exit → ``rc=0`` → dispatcher
 ``protocol_violation``.
@@ -17,7 +20,17 @@ import os
 from typing import Any, Iterable, Optional
 
 
-_TERMINAL_KANBAN_TOOLS = frozenset({"kanban_complete", "kanban_block"})
+_TERMINAL_KANBAN_TOOLS = frozenset({
+    "kanban_complete",
+    "kanban_block",
+    # request_review / request_changes are documented run-enders in the
+    # worker system prompt (KANBAN_GUIDANCE §5) and in the board state
+    # machine: request_review leaves the card in review (the dispatcher's
+    # clean-exit check only flags tasks still ``running``), and
+    # request_changes closes the review run and re-queues rework (#98107).
+    "kanban_request_review",
+    "kanban_request_changes",
+})
 
 _DEFAULT_MAX_ATTEMPTS = 2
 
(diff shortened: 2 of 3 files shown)


alt-glitch:

Duplicate of #97753 — both PRs implement the same terminal-handoff tool set in `agent/kanban_stop.py`; #97753 was opened first.

liuhao1024:
Agreed — the duplicate flag is correct. #97753 shipped the same fix (widening `_TERMINAL_KANBAN_TOOLS` with both handoff tools, same root cause, same tests shape) twelve hours before this PR opened; the issue I built against (#98107) has itself been marked a duplicate of the same report line. Closing this one in favor of #97753.

One observation from my investigation that #97753 may want to pick up, since it's the natural second half of the same bug: `build_kanban_stop_nudge`'s message text still teaches only two exits —

> Call `kanban_complete(summary=..., artifacts=[...])` if the work is done, OR `kanban_block(reason=...)` if you are blocked.

After the guard widening, a nudged build worker whose work is done-but-needs-review will read that as license to `kanban_complete` a card that should go through the review lane — the same wrong-exit the guard fix protects against, reintroduced through the nudge text. Updating the template to list all four run-enders (plus a line like "if a lifecycle call already succeeded, do not repeat it") closes that hole; I had pinned it locally with a `test_nudge_lists_all_run_enders` asserting all four tool names appear in the nudge.

Good catch on the ordering — deferring to #97753.

Issue #20270, opened 2026-05-05
[Feature]: Profile config inheritance from default


frap129:
### Problem or Use Case

Managing N profiles requires keeping N copies of config.yaml in sync. When I change my model or add an API key, I must edit all N profiles independently. This causes:

- Config drift — profiles diverge over time
- Duplication — same 200-line config copied N times
- Error-prone — easy to forget updating a profile

This is especially painful for Kanban workflows where you have multiple specialist profiles (researcher, engineer, reviewer, writer) that should share 95% of their config.

### Proposed Solution

Profiles should automatically inherit from `default` (~/.hermes/config.yaml). Profile config.yaml only needs to specify overrides:

~/.hermes/profiles/engineer/config.yaml
Inherits from default automatically — just specify what's different

model: anthropic/claude-opus-4
max_turns: 120

Resolution Order

1. Load default profile's config.yaml (~/.hermes/config.yaml)
2. Apply profile's overrides on top (deep merge for nested objects)
3. Profile-specific values win on conflicts

Example Use Case

~/.hermes/config.yaml (default / base)
model: anthropic/claude-sonnet-4
max_turns: 90
toolsets: [terminal, file, web, search]
mcp_servers:
  searxng:
    url: "https://search.example.com/mcp"
    http_auth_token: "${SEARXNG_TOKEN}"

~/.hermes/profiles/engineer/config.yaml (inherits automatically)
model: anthropic/claude-opus-4
max_turns: 120

~/.hermes/profiles/researcher/config.yaml (inherits automatically)
model: openai/o3-mini-high
toolsets: [web]

~/.hermes/profiles/sandbox/config.yaml (fully independent)
No inheritance — complete config from scratch
model: test-model
provider: local
max_turns: 50
... etc

CLI Creation Flag

hermes profile create engineer --inherits

This would create the profile with a skeleton config.yaml containing only comments explaining the inheritance behavior.

Implementation Notes

- Inheritance is automatic from default for all profiles
- Profile config.yaml contains only overrides — no special fields needed
- .env remains per-profile (API keys are already managed via hermes config set which routes to the right file)
- hermes config show should indicate inheritance chain (e.g., "Inherits from: default")

Scope: Medium (few files, < 300 lines)
Files affected:
- hermes_constants.py — get_hermes_home() + profile path resolution
- cli.py or hermes_cli/commands.py — hermes profile create logic
- Config loader (load_config()) — add inheritance merge step
[...]

PR #98115, opened 2026-08-29
feat(config): opt-in profile config inheritance via `inherit: true`

### LINKED ISSUE #98114: [Feature]: Opt-in profile config inheritance via `inherit: true`
### Problem

A named profile that omits a config key falls back to the built-in defaults, never to the root profile at `~/.hermes/config.yaml`. For most keys that isolation is exactly right. For `model` it produces a result that is hard to want: a profile whose `model:` block is absent resolves to `default: None` / `provider: None` rather than to the model the user is actually running.

The practical shape of this is a Bot Mode room. Five bot profiles, each a separate `HERMES_HOME`, all meant to run whatever model the user has chosen. Switching models means editing five files and getting `provider`, `default`, and `base_url` consistent in each — miss `provider` and the failure surfaces at the first API call, not at write time. Adding a sixth bot means remembering to do it again.

I hit this after moving a five-bot room from one provider to another; every profile needed the same three-key edit, and a profile I had "cleared" by deleting its `model:` block silently had no model at all.

### Prior discussion

#20270 asked for exactly this and was closed as won't-fix, with a rationale I think is correct as stated:

> Profiles are meant to be independent: each one owns its config, skills, and memory, and the default profile should never silently change a named profile's behavior at runtime.

The implementation in #20964 was reviewed as sound and closed for that design reason rather than any defect.

I am not asking to revisit that. The objection is to *implicit* inheritance — a coupling the profile owner never asked for. What is missing is a way for a profile to *declare* that it wants to follow the root, which leaves isolation as the default and makes the coupling visible in the profile's own config.

`hermes profile create --clone` covers the adjacent case (start from my setup, then diverge). It does not cover the fleet case, because a clone is a copy at creation time and the fleet needs to keep following.

### Proposal

An opt-in key in a **named** profile's config:

```yaml
# ~/.hermes/profiles/notion-expert/config.yaml
inherit: true
```

Semantics, chosen to keep the isolation model intact:

- **Off unless asked.** No key, or `inherit: false`, and behaviour is byte-for-byte what it is today.
- **Declared by the profile, not imposed by the root.** The root cannot opt a profile in; a reader of the profile's config can see the coupling in the file they are already looking at.
- **The profile always wins**, leaf by leaf. Setting `model.default` keeps the root's `model.base_url` instead of dropping the sibling keys.
- **No-op in the root profile**, which has no parent — not a loop, not an error.
- **Degrades to today's behaviour.** A root config that is missing or unparseable is skipped with a warning rather than taking a working profile down with it.

One implementation note that is easy to miss: the load cache keys on the config file's `(mtime, size)`. An inheriting profile depends on a second file that key does not watch, so without folding the root's signature into it, editing the root does not reach the profiles — which is precisely the operation the feature exists for. It fails silently and looks like the feature simply does not work.

### Scope

`hermes_cli/config.py` only — the merge is one layer between `DEFAULT_CONFIG` and the profile config, alongside the existing managed-scope layer. No new CLI surface, no changes to profile creation, no Python API change.

Happy to open a PR against this; I have the change working locally with tests covering opt-in, opt-out, per-leaf override, root-edit propagation, the root-profile no-op, and a corrupt root. If the direction is still unwanted, say so and I will drop it rather than re-litigate #20270.

andredezzy:
## What

Adds an opt-in `inherit: true` key that lets a **named** profile layer its config on top of the root profile's `~/.hermes/config.yaml`. Without the key, nothing changes.

Closes #98114.

## Why

A named profile that omits a key falls back to the built-in defaults, never to the root config. For most keys that isolation is right. For `model` it gives a result that is hard to want: a profile with no `model:` block resolves to `default: None` / `provider: None` rather than to the model the user is running.

The case that motivated this is a five-bot Bot Mode room. Every profile is its own `HERMES_HOME`, and switching models means editing five files with `provider`, `default`, and `base_url` consistent in each — miss `provider` and it fails at the first API call, not at write time.

## Relationship to #20270

#20270 asked for inheritance and was closed as won't-fix:

> Profiles are meant to be independent: each one owns its config, skills, and memory, and the default profile should never silently change a named profile's behavior at runtime.

That objection is to **implicit** inheritance — coupling the profile owner never asked for. This PR keeps it: isolation stays the default, and only a profile that declares `inherit: true` in its own config participates. The root cannot opt a profile in, so the coupling is always visible in the file a reader is already looking at.

`--clone` covers "start from my setup, then diverge". It does not cover a fleet that needs to keep following, because a clone is a copy at creation time.

If the direction is still unwanted, close it — I would rather drop this than re-litigate #20270.

## Design

- **Off unless asked.** No key, or `false`, and behaviour is what it is today.
- **Profile wins, leaf by leaf.** Overriding `model.default` keeps the root's `model.base_url` instead of dropping the siblings.
- **No-op in the root profile**, which has no parent — not a loop, not an error.
- **Degrades safely.** A missing or unparseable root config is skipped with a warning rather than taking a working profile down with it.

The merge is one layer between `DEFAULT_CONFIG` and the profile config, next to the existing managed-scope layer.

### The cache

Worth review attention. `_LOAD_CONFIG_CACHE` keys on the config file's `(mtime, size)`. An inheriting profile depends on a second file that key does not watch, so a root edit stayed invisible until something else touched the profile's own config — the fleet-wide switch this feature exists for would have silently not applied. The root signature is now folded into the cache key, matching what managed scope already does.

The root file is stat'd on every load rather than only when `inherit` is set: reading the key first would mean parsing the profile config on every load, which is the work the cache exists to avoid. A `stat()` per load is the cheaper end of that trade.

## Tests

`tests/hermes_cli/test_profile_inheritance.py`, 9 tests, written RED first:

- a profile without the key does not see the root model
- `inherit: true` pulls the root model and provider
- `inherit: false` is explicit isolation
- a profile value overrides the inherited one
- merging is per-leaf, not per-block
- editing the root reaches an inheriting profile (this one caught the cache bug)
- the root profile does not inherit from itself
- a missing root config leaves the profile working
- an unparseable root config does not break the profile

## Verification

- `pytest tests/hermes_cli/test_profile_inheritance.py` — 9 passed
- `pytest tests/hermes_cli/test_config.py` — 110 passed, 4 skipped
- `pytest tests/hermes_cli/test_managed_scope.py test_managed_installs.py` — 4 passed (shares the cache tuple this PR widens)
- `pytest tests/hermes_cli/` — 1055 passed, 1 failed

The one failure is `TestSanitizeEnvLines::test_migrate_reports_normalized_line_formatting`. It passes alone and **fails identically with this PR's changes stashed**, so it is pre-existing batch state, not this change. I have not run the full `tests/` suite end to end and am not claiming a green run I did not watch finish.


andredezzy:
Closing in favour of a branch cut from a clean `main` — this one was branched off my Bot Mode work and carried five unrelated desktop files into the diff. Reopening the same change, config-only, shortly.

alt-glitch:

Superseded by the clean active implementation in #98116; this closed predecessor also includes unrelated desktop bot changes.

PR #98116, opened 2026-08-29
feat(config): opt-in profile config inheritance via `inherit: true`

### LINKED ISSUE #98114: [Feature]: Opt-in profile config inheritance via `inherit: true`

andredezzy:
## What

Adds an opt-in `inherit: true` key that lets a **named** profile layer its config on top of the root profile's `~/.hermes/config.yaml`. Without the key, nothing changes.

Closes #98114.

## Why

A named profile that omits a key falls back to the built-in defaults, never to the root config. For most keys that isolation is right. For `model` it gives a result that is hard to want: a profile with no `model:` block resolves to `default: None` / `provider: None` rather than to the model the user is running.

The case that motivated this is a five-bot Bot Mode room. Every profile is its own `HERMES_HOME`, and switching models means editing five files with `provider`, `default`, and `base_url` consistent in each — miss `provider` and it fails at the first API call, not at write time.

## Relationship to #20270

#20270 asked for inheritance and was closed as won't-fix:

> Profiles are meant to be independent: each one owns its config, skills, and memory, and the default profile should never silently change a named profile's behavior at runtime.

That objection is to **implicit** inheritance — coupling the profile owner never asked for. This PR keeps it: isolation stays the default, and only a profile that declares `inherit: true` in its own config participates. The root cannot opt a profile in, so the coupling is always visible in the file a reader is already looking at.

`--clone` covers "start from my setup, then diverge". It does not cover a fleet that needs to keep following, because a clone is a copy at creation time.

If the direction is still unwanted, close it — I would rather drop this than re-litigate #20270.

## Design

- **Off unless asked.** No key, or `false`, and behaviour is what it is today.
- **Profile wins, leaf by leaf.** Overriding `model.default` keeps the root's `model.base_url` instead of dropping the siblings.
- **No-op in the root profile**, which has no parent — not a loop, not an error.
- **Degrades safely.** A missing or unparseable root config is skipped with a warning rather than taking a working profile down with it.

The merge is one layer between `DEFAULT_CONFIG` and the profile config, next to the existing managed-scope layer.

### The cache

Worth review attention. `_LOAD_CONFIG_CACHE` keys on the config file's `(mtime, size)`. An inheriting profile depends on a second file that key does not watch, so a root edit stayed invisible until something else touched the profile's own config — the fleet-wide switch this feature exists for would have silently not applied. The root signature is now folded into the cache key, matching what managed scope already does.

The root file is stat'd on every load rather than only when `inherit` is set: reading the key first would mean parsing the profile config on every load, which is the work the cache exists to avoid. A `stat()` per load is the cheaper end of that trade.

## Tests

`tests/hermes_cli/test_profile_inheritance.py`, 9 tests, written RED first:

- a profile without the key does not see the root model
- `inherit: true` pulls the root model and provider
- `inherit: false` is explicit isolation
- a profile value overrides the inherited one
- merging is per-leaf, not per-block
- editing the root reaches an inheriting profile (this one caught the cache bug)
- the root profile does not inherit from itself
- a missing root config leaves the profile working
- an unparseable root config does not break the profile

## Verification

- `pytest tests/hermes_cli/test_profile_inheritance.py` — 9 passed
- `pytest tests/hermes_cli/test_config.py` — 110 passed, 4 skipped
- `pytest tests/hermes_cli/test_managed_scope.py test_managed_installs.py` — 4 passed (shares the cache tuple this PR widens)
- `pytest tests/hermes_cli/` — 1055 passed, 1 failed

The one failure is `TestSanitizeEnvLines::test_migrate_reports_normalized_line_formatting`. It passes alone and **fails identically with this PR's changes stashed**, so it is pre-existing batch state, not this change. I have not run the full `tests/` suite end to end and am not claiming a green run I did not watch finish.


Enough1122:

**Overall:** Adds opt-in `inherit: true` for named profiles to layer `~/.hermes/profiles/<name>/config.yaml` on top of root `~/.hermes/config.yaml` (leaf-level deep merge). Fixes silent empty-model fallback to provider's first catalog entry for fleet-managed installs. Behind explicit opt-in, backward compatible.

**What it does**
- `hermes_cli/config.py` adds `_resolve_root_config_path` (returns `None` for root profile, else `<root>/profiles/<name>/../..` logic via `parent.parent.name=="profiles"` check), `_load_inherited_config` (reads root, warns via `_warn_config_parse_failure` on failure, returns `{}`), and merges in `_load_config_impl`: `config = _deep_merge(config, inherited)` then `_deep_merge(config, user_config)` so profile keys win leaf-wise (overriding `model.default` keeps root's `model.base_url`).
- Extends `_LOAD_CONFIG_CACHE` from 4-tuple to 6-tuple `(user_mtime,size, managed_mtime,size, root_mtime,size)` and stat's `root_cfg_path` unconditionally each load — avoids parsing profile to check `inherit` (which is the work the cache avoids). Cache check compares `cached[:6]==cache_sig`; LKG path also updated.
- `cli.py:load_cli_config` mirrors same layering for the TUI's direct file reader (which bypasses `load_config()`), ensuring the agent model matches the config file.
- Docs in `cli-config.yaml.example` and tests: `test_cli_config_inheritance.py` (subprocess isolation for `CLI_CONFIG` at import time) and `test_profile_inheritance.py` (opt-in isolation, leaf merge, root-edit invalidation, root self-inherit no-op, missing/broken root graceful).

**Non-blocking notes**
- `root_sig` is stat'd unconditionally even for non-inheriting profiles — adds one `stat()` per `load_config()` for every profile. Negligible but worth noting; avoids per-load YAML parse of profile to check `inherit` flag, which is more expensive.
- `_resolve_root_config_path` checks `profile_dir.parent.name != "profiles"` — if `_get_profiles_root` is ever custom or nested deeper, this heuristic could mis-route. Assuming canonical layout `profiles/<name>/config.yaml`.
- `_load_inherited_config` returns raw YAML dict, then caller does `_normalize_root_model_keys` — ordering is `inherited` normalized separately from `file_config`/`user_config`, which is correct; ensure `inherit` key itself is popped after normalization so it doesn't leak into merged config (it is popped).
- Editing root while profile is inheriting correctly invalidates cache via `root_sig`; editing root when profile moved from `inherit:false` to `true` also picks up because root sig changed — good.

Clean opt-in with good cache discipline.

andredezzy:
Good question, and the honest answer was worse than I expected — so it is fixed in `a3057a6`.

I checked it against the real loader rather than reasoning about it. A profile whose config is:

```yaml
model:
  inherit: true
```

resolves to `model.default = None`. The key is read only at the top level, so a nested one is silently discarded, and the section it was meant to fill stays empty. The user then meets a missing model with no hint that a misplaced line caused it — the worst shape a config mistake can take.

**Why not make it work per-section.** That would give the same idea two spellings, and a half-inherited file has no obvious merge order: if `model.inherit` is true but `agent` is absent, does `agent` inherit? Either answer surprises someone. Inheritance stays a whole-file switch.

**What changed instead.** A profile that does not inherit is now checked for a nested `inherit`, and each occurrence is named on stderr:

```
⚠ /path/to/config.yaml: model.inherit has no effect — `inherit: true` applies to the whole file and must sit at the top level.
```

Two tests cover it: one asserting the misplaced key is reported, one asserting the correct top-level spelling stays silent. `test_profile_inheritance.py` + `test_cli_config_inheritance.py` are 14 passed.

Thanks for catching it — the silent path was the actual bug here.

andredezzy:
Thanks — went through the non-blocking notes properly rather than filing them as acknowledged. One of them was hiding a real defect.

**`_resolve_root_config_path` and the canonical layout.** This one was right, and worse than "could mis-route". I tested both cases against the loader:

- A custom `HERMES_HOME` outside `~/.hermes` (the Docker case) inherits **correctly** — resolution is relative to the config path, not a hardcoded home. `model` and `agent.max_turns` both come through.
- A profile nested deeper, `profiles/team/member/`, does **not**. The lookup walks a fixed two levels up, finds no root, and `inherit: true` silently does nothing. The profile ends up with `model = ''` and nothing on screen explaining why.

That is precisely the failure mode this feature exists to remove, so `f467940` makes it speak: an `inherit: true` that cannot reach a root now names the expected layout on stderr, next to the nested-`inherit` warning added in `a3057a6`.

I left the layout requirement itself alone. Searching upward for a root would let a profile inherit from an unrelated ancestor directory — a worse surprise than a warning.

**The unconditional `stat()`.** Measured rather than assumed, on this machine:

```
stat()               1.17 µs
load_config() warm  360.90 µs
fraction              0.3 %
```

Your reasoning holds — it is cheaper than parsing the profile YAML to decide whether the stat is needed. Leaving it as is.

**`inherit` leaking into the merged config.** Confirmed popped, and the nested-`inherit` test would catch a regression there.

Full state: `test_profile_inheritance.py` + `test_cli_config_inheritance.py` 15 passed. Both new behaviours verified RED→GREEN — with the warning removed, `test_an_unreachable_root_is_reported_not_swallowed` fails.

andredezzy:
Merged current upstream into this branch and resolved the config-loader conflicts against its extracted helpers. The diff remains limited to profile inheritance, its documentation, and tests.

Verification on `a1fc4f2727`: 137 Python tests passed, 4 platform-specific tests skipped; Ruff and Python compilation passed. The root-file cache invalidation and the existing nested/unreachable-inheritance warnings remain covered. GitHub now reports the branch mergeable.

xaviergmail:
Tested and confirmed working for my use case (multiple profiles inheriting default profile)
The one caveat I found was updates to the default profile (performed via the desktop app, gateway is remote, multiplexing enabled) did not propagate to other profiles without restarting the gateway. But that's a minor inconvenience that can be addressed later on, and the benefit from this PR alone is a huge QoL improvement.

teknium1:
Thanks @andredezzy for the careful loader work here, and to @xaviergmail for testing it.

We're not adding live config inheritance between profiles. Profiles are isolated islands by design: each profile's `config.yaml` / `.env` / state is complete on its own so a change in one can never alter another's behaviour — the same reason #20270 was declined. "Start from the default profile" is covered by `hermes profile create --clone` (copies `config.yaml`, `.env`, `SOUL.md` and curated memory at creation time). The caveat you found — changes to the root config not propagating until the gateway restarts — is the kind of cross-profile coupling this design avoids.

Closing as a design decision, not a quality judgement on the implementation.

Issue #68474, opened 2026-07-21
state.db zeroed (95MB of null bytes) during desktop update to v0.19.0 on Windows


Pofium:
## Bug Description

After updating the Hermes desktop app to v0.19.0 on Windows 10, the `state.db` file (SQLite session database) was completely destroyed — all 99,913,728 bytes replaced with null bytes (`\x00`). The pre-update snapshot (`state-snapshots/20260720-063356-pre-update/state.db`) was intact and valid.

## Steps to Reproduce

1. Have an existing `state.db` with sessions (50 sessions, 13,089 messages in this case)
2. Update desktop app from previous version to v0.19.0
3. Observe `state.db` is now all zero bytes

## Expected Behavior

`state.db` should survive the update process intact, or be properly migrated if the schema changed.

## Actual Behavior

`state.db` became 95MB of null bytes. The SQLite header (`SQLite format 3\0`) was completely gone. The WAL file (`state.db-wal`) was also emptied to 0 bytes.

## Environment

- OS: Windows 10
- Version: 0.19.0 (desktop build: 2026-07-21T07:26:04Z, contentHash: 0ced1722...)
- Update exit code: 0 (reported success!)
- Pre-update version: unknown (snapshot dated 2026-07-20 06:33:56 MSK)
- Python: 3.11.15 (venv), 3.14.3 (system)

## Recovery

Restored from pre-update snapshot at `state-snapshots/20260720-063356-pre-update/state.db` which was valid SQLite. Sessions between the snapshot (July 20 06:33) and the update (July 21 10:26 MSK) were lost.

The zeroed file is preserved at `state.db.zeroed.bak`.

## Diagnostic Output

```
$ file state.db
state.db: data   (NOT "SQLite 3.x database")

$ xxd state.db | head
00000000: 0000 0000 0000 0000 0000 0000 0000 0000
(all zeros for entire 95MB)

$ python -c "import sqlite3; sqlite3.connect(state.db)"
sqlite3.DatabaseError: file is not a database
```

Update metadata:
```json
{"ts": 1784619119, "behind": 0, "rev": null, "ver": "0.19.0"}
// update_exit_code: 0
```

## Additional Context

- The pre-update snapshot mechanism worked correctly — the snapshot was valid SQLite
- The update process reported exit code 0 (success), masking the data loss
- This is a data-loss bug: sessions between the snapshot and the update are unrecoverable
[...]

PR #98121, opened 2026-08-29
fix(desktop): snapshot every profile state.db before update (#97994)

### LINKED ISSUE #97994: [Bug]: pre-update state.db guard (#68474) covers only the root DB — profile databases get no emergency backup
### Summary

The pre-update `state.db` integrity guard added for #68474 protects **only the root `$HERMES_HOME/state.db`**. It never iterates `profiles/*/state.db`, so on a multi-profile install every profile database goes through the update with no emergency backup and no header check.

On a 13-database install (root + 12 profiles) that is **1 of 13 databases protected**.

During a `v0.20.6` desktop self-update on macOS this cost me a 150 MB profile database. It survived only because a dying backend process still held an open file descriptor to the unlinked inode, which I recovered with `lldb`. There was no other copy.

### The code

`apps/desktop/electron/main.ts` (current `main`, and identical in the shipped 0.20.6):

```js
// ── Pre-flight state.db integrity guard (#68474) ─────────────────────
// Take an emergency snapshot of state.db and verify the live copy is
// intact before any update process mutates the install.
function preflightStateDb(hermesHome, rememberLog) {
  const stateDbPath = path.join(hermesHome, 'state.db')
  ...
}
```

Both call sites pass the root home and nothing else:

```
3737:    preflightStateDb(HERMES_HOME, rememberLog)
4202:    preflightStateDb(HERMES_HOME, rememberLog)
```

`grep profiles` inside the function body returns nothing. The same file references `profiles` 65 times elsewhere, so the multi-profile layout is well known to this module — the guard simply predates it or was never extended.

Directly observable consequence on my install: the root home holds three `state.db.pre-update-emergency-*.bak` files, and **not one profile directory holds any**.

### What happened

Desktop self-update on macOS 15.6 (Apple Silicon), Hermes 0.20.6, 12 profiles. From `~/.hermes/logs/desktop.log` (times local, profile anonymised as `profile-A`):

```
16:29:08  [updates] state.db pre-flight: size=170627072, headerOk=true      <- root only
16:29:08  [updates] emergency state.db backup: ...pre-update-emergency-...bak
16:29:11  Hermes backend for profile "profile-A" exited (SIGTERM)
          (…all 12 profile backends SIGTERM'd; none backed up…)
16:31:31  Starting Hermes backend for profile "profile-A"
16:36:05  Hermes backend for profile "profile-A" exited (SIGKILL)
16:36:11  Hermes backend for profile "profile-A" exited (SIGKILL)
16:37:21  Hermes backend for profile "profile-A" exited (SIGKILL)
16:39:53  Hermes backend for profile "profile-A" exited (SIGKILL)
16:40:04  Hermes backend for profile "profile-A" exited (SIGKILL)   <- 5 hard kills
16:40:04  Starting Hermes backend for profile "profile-A"           <- pid P opens the real 150MB db
16:40:58  *** a fresh, empty state.db is born at the profile's path ***
16:41:07  hermes_state: state.db write failed with an FTS-corruption error
          (database disk image is malformed) — attempting one-shot in-place FTS rebuild
16:41:07  hermes_state: In-place FTS rebuild made no progress; the database needs
          the full offline repair path (repair_state_db_schema).
16:41:07  hermes_state: Could not detach corrupt FTS indexes; canonical write
          still cannot proceed: database disk image is malformed
```

`stat` on the replacement file gives birth time `16:40:58`, a new inode, and a valid `SQLite format 3` header — i.e. SQLite created it fresh on an empty path. The original 150 MB inode was gone from the filesystem: `find -inum` located it nowhere, and `lsof` showed it held open only by the backend started at 16:40:04.

**The original database was never actually corrupt.** After recovering it byte-exact from the live fd, it passed `PRAGMA integrity_check`, `PRAGMA quick_check`, and an `fts5` `integrity-check` on both FTS tables, with all 409 sessions / 8,907 messages intact and current to 16:38:53 — three minutes before the incident. The `malformed` errors at 16:41:07 are the running process's view *after* its file was swapped underneath it (it retained the old main-db inode while the `-wal`/`-shm` at the path became new inodes), not damage to the data.

### What I could not determine

I could not identify the call that removed the original file. Nothing logs it. It was **not** `quarantine_zeroed_state_db` — that renames to `state.db.zeroed-<ts>-<pid>.bak` and logs; no such file exists and "zeroed" appears in no log. It was **not** `repair_state_db_schema` — that is explicitly inode-preserving (`_copy_database_snapshot` into the existing path, "Do not os.replace the live DB"), it has a non-destructive test suite, and the log shows it was only *recommended*, never invoked.

So there appears to be a second, unlogged path that can remove a profile `state.db`, reachable when a backend is hard-killed repeatedly around an update. The 5 consecutive `SIGKILL`s are themselves worth a look — whatever watchdog issues them is the entry condition.

Either way, the guard gap stands on its own: **had `preflightStateDb` covered profiles, an emergency copy would have existed and none of the above would have mattered.**

### Suggested fix

Extend the guard to every database it is meant to protect:

```js
function preflightAllStateDbs(hermesHome, rememberLog) {
  preflightStateDb(hermesHome, rememberLog)                       // root, unchanged
  const profilesDir = path.join(hermesHome, 'profiles')
  if (!fs.existsSync(profilesDir)) return
  for (const name of fs.readdirSync(profilesDir)) {
    const home = path.join(profilesDir, name)
    if (fs.statSync(home).isDirectory()) preflightStateDb(home, rememberLog)
  }
}
```

`preflightStateDb` already takes a home directory as its parameter, so it works unmodified per profile. Both call sites switch to `preflightAllStateDbs`.

Worth pairing with a disk-space guard: on a large install this multiplies the emergency-backup footprint by the profile count (mine is ~1.1 GB across 13 databases), and #75724 shows the backup path is already sensitive to what else lives in `HERMES_HOME`. Capping or pruning per-profile emergency copies the way the root one is pruned to 2 would keep that bounded.

### Related

- #68474 — the original zeroed-`state.db`-during-update bug this guard was added for (closed). This report is that fix having only ever covered 1 of N databases.
- #88252 — FTS5 corruption reported after every update, emergency `state.db` handling (open, Windows).
- #75724 — full pre-update backup aborts on a non-SQLite `.db` in `HERMES_HOME` (open).

### Environment

- Hermes 0.20.6, desktop self-update
- macOS 15.6, Apple Silicon
- 13 SQLite databases (root + 12 profiles), ~1.1 GB total
- Verified against current `main`: the `preflightStateDb` block is byte-identical to the shipped copy and still root-only.

smfworks:
## Summary
- Pre-update \`state.db\` guard (#68474) only covered the root home. Profile databases under \`profiles/*/state.db\` had no emergency backup.
- \`preflightStateDb\` now snapshots every \`profiles/<name>/\` directory as well as the root. Missing profile DBs are skipped without failing the root backup.

Closes #97994

## Test plan
- [x] \`npx vitest run electron/preflight-state-db.test.ts\` — 3 passed

## Verification
All six adversarial checks PASS. Python-level \`hermes update\` snapshot is a sibling path, left unchanged.

### DIFF
--- a/apps/desktop/electron/main.ts
+++ b/apps/desktop/electron/main.ts
@@ -279,6 +279,7 @@ import { selectPoolEvictions } from './pool-eviction'
 import { createPoolStopper } from './pool-stop'
 import { poolTouchKeys } from './pool-touch-scope'
 import { createKeepAwake } from './power-save'
+import { preflightStateDb } from './preflight-state-db'
 import { PreviewReachRegistry } from './preview-reach'
 import {
   createPrimaryRemoteConnection,
@@ -4085,91 +4086,14 @@ function runningAppBundle() {
   return dir.endsWith('.app') ? dir : null
 }
 
-// ── Pre-flight state.db integrity guard (#68474) ─────────────────────
-// Take an emergency snapshot of state.db and verify the live copy is
-// intact before any update process mutates the install.  Runs in the
-// desktop Electron process itself, before the backend is killed and
+// ── Pre-flight state.db integrity guard (#68474, #97994) ─────────────────────
+// Snapshot every state.db (root + profiles/*) and verify the live copies
+// look like SQLite before any update process mutates the install. Runs in
+// the desktop Electron process itself, before the backend is killed and
 // before the updater is spawned — a separate safety net from the
 // Python-level pre-update snapshot inside `hermes update`.
-function preflightStateDb(hermesHome, rememberLog) {
-  const stateDbPath = path.join(hermesHome, 'state.db')
+// Implementation: ./preflight-state-db.ts
 
-  if (!fileExists(stateDbPath)) {
-    rememberLog('[updates] state.db pre-flight: not found (fresh install?)')
-
-    return
-  }
-
-  try {
-    const stat = fs.statSync(stateDbPath)
-
-    if (stat.size > 100) {
-      const fd = fs.openSync(stateDbPath, 'r')
-      const header = Buffer.alloc(16)
-
-      fs.readSync(fd, header, 0, 16, 0)
-      fs.closeSync(fd)
-
-      const expectedHeader = Buffer.from('SQLite format 3\0')
-      const headerOk = header.equals(expectedHeader)
-
-      rememberLog(
-        `[updates] state.db pre-flight: size=${stat.size}, ` +
-          `headerOk=${headerOk}, headerHex=${header.toString('hex')}`
-      )
-
-      if (!headerOk) {
-        rememberLog(
-          '[updates] state.db header is INVALID before update — ' +
-            'this indicates pre-existing corruption or a concurrent write issue'
-        )
-      }
-
-      // Emergency timestamped backup, separate from the Python-level snapshot.
-      const ts = new Date().toISOString().replace(/[:.]/g, '-')
-
-      const emergencyPath = path.join(hermesHome, `state.db.pre-update-emergency-${ts}.bak`)
-
-      try {
-        fs.copyFileSync(stateDbPath, emergencyPath)
-        const emergStat = fs.statSync(emergencyPath)
-
-        rememberLog(`[updates] emergency state.db backup: ${emergencyPath} ` + `(${emergStat.size} bytes)`)
-
-        // Prune to the 2 most recent emergency backups.
-        try {
-          const homeDir = fs.readdirSync(hermesHome)
-
-          const backups = homeDir
-            .filter(
-              f =>
-                f.startsWith('state.db.pre-update-emergency-') &&
-                f.endsWith('.bak') &&
-                f !== path.basename(emergencyPath)
-            )
-            .sort()
-            .reverse()
-
-          for (const old of backups.slice(2)) {
-            try {
-              fs.unlinkSync(path.join(hermesHome, old))
-            } catch {
-              void 0
-            }
-          }
-        } catch {
-          void 0
-        }
-      } catch (copyErr) {
-        rememberLog(`[updates] emergency state.db backup failed: ${copyErr.message}`)
-      }
-    } else {
-      rememberLog(`[updates] state.db too small (${stat.size} bytes) for a valid SQLite database`)
-    }
-  } catch (statErr) {
-    rememberLog(`[updates] could not stat state.db before update: ${statErr.message}`)
-  }
-}
 
 // macOS/Linux update hand-off: spawn the repo-owned posix orchestrator
 // (scripts/desktop-update/posix.sh) detached and QUIT. The script waits us
--- a/apps/desktop/electron/preflight-state-db.test.ts
+++ b/apps/desktop/electron/preflight-state-db.test.ts
@@ -1,0 +1,85 @@
+import fs from 'node:fs'
+import os from 'node:os'
+import path from 'node:path'
+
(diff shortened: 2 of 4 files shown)


Enough1122:

**Overall:** Expands pre-update `state.db` guard to cover `profiles/*/state.db` (#97994). Previously only root home had an emergency snapshot; profile DBs were unprotected. Extracts logic to a typed module with tests.

**What it does**
- Extracts `preflightStateDb` from `apps/desktop/electron/main.ts` (≈78 lines deleted) into `apps/desktop/electron/preflight-state-db.ts` with `listStateDbHomes(hermesHome)` (root + sorted `profiles/*` directories, file entries ignored) and `preflightOneStateDb(dbHome, rememberLog)` (header check, `MIN_SQLITE_BYTES=100`, timestamped `state.db.pre-update-emergency-<iso>.bak` copy, prune to 2 most recent per-home).
- `main.ts` now `import { preflightStateDb } from './preflight-state-db'` and wrapper comment retained.
- Tests `preflight-state-db.test.ts` cover: lists root + every profile directory (sorted, ignores file), snapshots profile DBs alongside root (checks `emergencyBackups` count 1 per home and log contains `profiles`), and missing profile DB skipped without failing root backup.
- Logging now includes `path=` and per-home emergency path, aiding diagnosis.

**Non-blocking notes**
- `listStateDbHomes` sorts entries — deterministic order for logs and backups, good.
- `fs.existsSync` + `fs.statSync` in `preflightOneStateDb` is TOCTOU-safe enough for pre-flight (no concurrent writer expected before updater spawn), but a disappearing `state.db` between checks is caught by outer try/catch.
- Prune keeps 2 most recent per `dbHome` (`slice(2)` after reverse sort by filename which is ISO timestamp — works because ISO lexicographic equals chronological). Consider using mtime if clock skew, but filename ordering is fine.
- No change to Python-level `hermes update` snapshot — this is the Electron-side safety net before backend is killed, complementary.

Well-scoped reliability.

teknium1:
Reviewed against current `main` as part of the profiles/multiplex sweep — this correctly covers #97994.

- Verified the premise still holds on main: the pre-update guard (#68474 lineage) snapshots only the root home's `state.db`; `profiles/*/state.db` get no emergency backup.
- The extraction of the inline `preflightStateDb` into `preflight-state-db.ts` preserves root-DB behavior byte-for-byte, and `listStateDbHomes` enumerates root + `profiles/*` tolerantly (missing/unreadable profiles dir doesn't throw). Per-home isolation means one failing profile DB doesn't block the others' backups. Vitest coverage is real (backup created per home, empty-profiles skip) where the original inline code had none.

Two pre-existing nits carried over faithfully (fine to keep out of scope, noting for the record):
1. `fs.existsSync` probe is TOCTOU-y — was already true for the root path.
2. The prune keeps 3 backups per home, not 2 (prune runs before the new backup is written, then `slice(2)`) — same off-by-one as the original.

Looks merge-ready pending maintainer review.

teknium1:
Thanks for the work here @smfworks, but we're not taking this direction — maintainer call. Copying every profile's state.db before each update multiplies pre-update I/O and disk usage with the number of profiles (these DBs can be multi-GB), and the update path needs to get leaner, not heavier. The 0-byte quarantine + corruption detection work that landed this week (#99513, #98017) addresses the actual loss modes more surgically. Appreciate the effort — the referenced issue stays open for a lighter-weight approach.

PR #19007, opened 2026-05-02
fix(compression): use extract_content_or_reasoning for thinking models

shellybotmoyer:
## Summary

Fixes #19003

Context compressor ignores the `reasoning` field when extracting summary content from the auxiliary LLM response. When using thinking/reasoning models (DeepSeek v4, GLM-5.1, Qwen 3.5, etc.) via Ollama 0.22+, these models return their output in the `reasoning` field with `content` as an empty string — especially when `max_tokens` is constrained (which compression does). The compressor gets an empty summary and falls back to a static context marker, losing the entire compaction.

## Bug

`agent/context_compressor.py` line 871:
```python
content = response.choices[0].message.content  # BUG: ignores reasoning field
```

When a thinking model returns `content=""` and `reasoning="..."`, this line produces an empty string. The compressor then falls back to: `"Summary generation failed — inserting static fallback context marker"`

In our deployment, 3 of 25 compression events (12%) produced fallback markers.

## Fix

Replace raw `.message.content` access with the existing `extract_content_or_reasoning()` helper from `auxiliary_client.py`, which checks `content` first, then falls back to `reasoning`, `reasoning_content`, and `reasoning_details`.

```python
from agent.auxiliary_client import call_llm, extract_content_or_reasoning
# ...
content = extract_content_or_reasoning(response)
```

The `extract_content_or_reasoning()` function already exists and is used by the main agent loop — this just applies the same reasoning-field handling to the compression path.

## Testing

- Tested with Ollama 0.22.1+ and `deepseek-v4-flash:cloud` as compression auxiliary
- Before fix: `content=""`, `reasoning` held the actual summary → fallback marker inserted
- After fix: `extract_content_or_reasoning()` returns the reasoning content → proper summary

## Impact

- One-line change + one import addition
- No behavior change for models that return content normally (the common path)
- Fixes 12% compression failure rate with thinking models

Issue #19003, opened 2026-05-02
[Bug] Context compressor ignores reasoning field — empty summaries with thinking models (Ollama 0.22+)


mikronn2:
### Summary

`context_compressor.py` reads only `response.choices[0].message.content` (line 871) when extracting the summary from the auxiliary LLM. When using thinking/reasoning models (DeepSeek v4, GLM-5.1, Qwen 3, etc.) via Ollama 0.22+, these models return their output in the `reasoning` field with `content` as an empty string — especially when `max_tokens` is constrained (which compression does). The compressor gets an empty summary and falls back to a static context marker, losing the entire compaction.

### Root Cause

Ollama 0.22.x changed how thinking models return responses. The `reasoning` field is now always populated for models with a thinking renderer (DeepSeek v4, GLM-5.1, Qwen 3.5, etc.), even without `think: true`. With limited `max_tokens`, the model's reasoning tokens consume the entire budget, and `content` comes back empty.

Hermes already has `extract_content_or_reasoning()` in `auxiliary_client.py` (line 3561) that handles this exact case — it checks `content` first, then falls back to `reasoning`, `reasoning_content`, and `reasoning_details`. But `context_compressor.py` doesn't use it.

### Affected Code

`agent/context_compressor.py` line 871:
```python
content = response.choices[0].message.content  # BUG: ignores reasoning field
```

Should be:
```python
from agent.auxiliary_client import call_llm, extract_content_or_reasoning
# ...
content = extract_content_or_reasoning(response)
```

### Reproduction

1. Configure Hermes with Ollama 0.22.1+ and a thinking model (e.g., `deepseek-v4-flash:cloud` or `glm-5.1:cloud`) as the compression auxiliary
2. Have a conversation long enough to trigger context compression
3. Observe that the compression summary is empty — the model's output went entirely to `reasoning` while `content` is `""`
4. Compressor falls back to a static context marker: "Summary generation failed — inserting static fallback context marker"

### Evidence

In our deployment, 3 of 25 compression events produced fallback markers (12% failure rate). After upgrading to Ollama 0.22.1, testing with `deepseek-v4-flash:cloud` and `max_tokens: 100` returns:
- `content: ""` (empty)
- `reasoning: "..."` (928 chars of actual summary content)

With `max_tokens: 300`, content returns 75 chars while reasoning takes 929 chars — the reasoning field holds the actual compressed content the compressor needs.

### Impact

[...]

PR #98125, opened 2026-08-29
fix(compression,title): use extract_content_or_reasoning for thinking models

shellybotmoyer:
## Summary

Two remaining call sites still used raw `response.choices[0].message.content` instead of the existing `extract_content_or_reasoning()` helper:

1. **`agent/context_compressor.py`** — the lean digest path (`_LEAN_DIGEST_PROMPT`) at the `resp.choices[0].message.content` site. This is a **different code path** from the one fixed in PR #19007 / commit `b0f71445c4`, which fixed the main summary compression path. The lean digest fallback was missed.
2. **`agent/title_generator.py`** — the `generate_title()` function reads `response.choices[0].message.content` directly. With reasoning models, this returns empty string, producing empty session titles.

## Bug Class

Same class as issue #19003: thinking/reasoning models (DeepSeek v4, GLM-5.1, Qwen 3 via Ollama 0.22+) put their output in `message.reasoning` / `message.reasoning_content` fields instead of `message.content`. The existing `extract_content_or_reasoning()` helper in `agent/auxiliary_client.py` handles all three fallback paths (content → reasoning → reasoning_details).

## Changes

- `agent/context_compressor.py`: import `extract_content_or_reasoning`, replace `resp.choices[0].message.content` with `extract_content_or_reasoning(resp)` in the lean digest path
- `agent/title_generator.py`: import `extract_content_or_reasoning`, replace `response.choices[0].message.content` with `extract_content_or_reasoning(response)`

## Test Plan

- [ ] Existing compression tests pass (`pytest tests/test_context_compressor.py`)
- [ ] Existing title generation tests pass (`pytest tests/test_title_generator.py`)
- [ ] With a reasoning model (e.g., DeepSeek v4 via Ollama), session titles are no longer empty
- [ ] Lean digest compression produces non-empty output with reasoning models


Enough1122:

**Overall:** Fixes lean-digest and title paths for thinking models that expose reasoning in `reasoning_content` rather than `content`. Follows existing `extract_content_or_reasoning` helper pattern.

**What it does**
- `agent/context_compressor.py:_build_chunk_digests` changes `resp.choices[0].message.content` → `extract_content_or_reasoning(resp)` (with `hasattr(resp,"choices")` guard retained). This lean-digest path (`_LEAN_DIGEST_PROMPT`) was separate from the main summary fix in #19007 and was missed.
- `agent/title_generator.py:generate_title` same replacement (`response.choices[0].message.content` → `extract_content_or_reasoning(response)`), imports helper from `agent.auxiliary_client`.

**Non-blocking notes**
- `extract_content_or_reasoning` already handles `None`/empty and `reasoning_content` fallback — correct for thinking models (e.g., Claude extended thinking, DeepSeek). No behavior change for non-thinking models.
- Both call sites retain `or ""` downstream, so `None` return is coalesced to empty string — safe; `strip_think_blocks` / `_clean_title` handle empty gracefully (likely skip title).
- Consider adding a small test pinning that a mock thinking response (`content=None, reasoning_content="summary"`) returns summary, to prevent future regression on these specific call sites.

Trivial, correct.

teknium1:
Heads-up: PR #98628 (merged, 4f22543509) removed the lean digest call site this PR patches in agent/context_compressor.py — that half is now moot (the single summary call site may still warrant the same extract_content_or_reasoning treatment, worth checking). The agent/title_generator.py half looks still valid on current main. Recommend rebasing to drop the digest hunk; the title fix can then be reviewed standalone.

PR #98127, opened 2026-08-29
feat(approvals): confirm gate for agent-initiated model-routing config writes

### LINKED ISSUE #97652: [Feature]: Confirm gate for agent-initiated changes to model-routing config

DavidMetcalfe:
## What does this PR do?

Adds a confirm gate for agent-initiated writes to model-routing configuration
keys (`model.*`, `delegation.model/provider/base_url`, bare
`delegation`/`auxiliary` section writes, `auxiliary.<task>.model/provider/base_url`).

Hermes agents can rewrite this configuration via a single
`hermes config set` — silently, even when the user runs `approvals.mode: off`.
The write determines per-token cost for every subagent, so a silent rewrite is
an unreviewed billing decision. A real incident motivated this: an agent
session pattern-matched the user's *valid* `delegation.model` against a stale
memory of a different, hyphenated, invalid slug and overwrote it with zero
errors occurring; the user discovered the change days later in Settings.

The gate is modeled on the existing `approvals.destructive_slash_confirm`
pattern — a confirmation class that survives `approvals.mode: off` with its
own opt-out key.

## Design

- **Discrimination**: the agent command-execution layer stamps every spawned
  subprocess env with `HERMES_AGENT_INITIATED=1` (stamped in
  `tools/environments/base.py`, so local, docker, ssh, modal, daytona, and
  vercel_sandbox backends are all covered). A human's own
  `hermes config set` in their shell never carries the marker and is never
  gated.
- **Agent-marked writes to model-routing keys**: refused (non-zero exit, clear
  stderr message) — never prompted. An agent process must not block on stdin
  (PTY-backed terminal runs have a real TTY; prompting would hang the tool
  call for its full timeout). The refusal instructs the agent to surface the
  proposed change to the user.
- **User-directed writes**: `hermes config set <key> <value>
  --confirm-model-change` marks the write as user-directed, bypasses the
  refusal, and emits a visible one-line change notice
  (`⚠ key changed: old → new`).
- **Gate opt-out**: `approvals.model_config_confirm: false` (default `true`)
  downgrades the gate to the visible change notice.
- **Secret hygiene**: values shown in notices/refusals pass through
  `_gate_redacted()`, which masks credential-shaped leaf keys (including
  nested values in `--force` section writes).
- Fail-open: any detection error ungates the write (a human write is never
  mis-gated).

## Changes Made

- `hermes_cli/config.py` — `_is_model_routing_key`, `_is_agent_initiated_write`,
  `_maybe_gate_model_routing_write`, `_gate_redacted`, notice/refusal emitters;
  `set_config_value` gains `confirm_model_change` and the gate call.
- `tools/environments/base.py` + per-backend wiring — the env stamp.
- `hermes_cli/config_defaults.py` — `approvals.model_config_confirm: True` with
  full documentation comment.
- `hermes_cli/subcommands/config.py` — `--confirm-model-change` CLI flag.
- `tests/hermes_cli/test_model_config_confirm_gate.py` — 45 tests (was 30 after
  round 1, extended through the review loop).

## How to Test

1. `hermes config set delegation.model anything` from your own shell → writes
   normally (no marker).
2. Same command with `HERMES_AGENT_INITIATED=1` env → refused, non-zero exit,
   no hang.
3. Same with `--confirm-model-change` → allowed, change notice printed.
4. `approvals.model_config_confirm: false` → agent-marked writes allowed with
   notice.

## Review

Two rounds of cross-vendor review (Gemini 3.7 Flash High + GPT-OSS 120B Medium,
byte-identical briefs with the full diff; verified-against-source dispositions
for every finding, including rejected misreads, are in the commit bodies).
Round 2 verdicts: all ACCEPTABLE across both reviewers; remaining SHOULD-FIXes
(nested-secret redaction) applied and tested.

Closes #97652

### DIFF
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -5667,7 +5667,7 @@ def _lookup_old_value(key: str):
         return None
 
 
-def _gate_redacted(value, key) -> str:
+def _gate_redacted(value, key):
     """Mask credential-shaped values so the gate notice/refusal never leaks a secret.
 
     ``_is_model_routing_key`` matches ``model.*``, which includes
@@ -5678,6 +5678,14 @@ def _gate_redacted(value, key) -> str:
     value whose leaf key is credential-shaped is redacted for display.
     """
     leaf = key.rsplit(".", 1)[-1].lower()
+    if isinstance(value, dict):
+        # Section writes (--force delegation / auxiliary) can carry nested
+        # credential-shaped leaves; sanitize recursively (Flash R2 SHOULD-FIX).
+        return {
+            k: _gate_redacted(v, f"{key}.{k}") if isinstance(v, (dict, list))
+            else (_gate_redacted(v, f"{key}.{k}") if isinstance(v, str) and v else v)
+            for k, v in value.items()
+        }
     if leaf in _SECRET_CONFIG_KEYS and isinstance(value, str) and value:
         from agent.redact import mask_secret
 
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -5567,7 +5567,165 @@ def _coerce_float(value: str):
     return f
 
 
-def set_config_value(key: str, value: str, force: bool = False):
+# ---------------------------------------------------------------------------
+# Model-routing config confirm gate (#97652)
+# ---------------------------------------------------------------------------
+# Agents can silently rewrite the model-routing config (``model.*``,
+# ``delegation.model/provider/base_url``, ``auxiliary.*.model/provider/base_url``)
+# via a single ``hermes config set`` — even with ``approvals.mode: off``.  That
+# determines what the user pays per token for every subagent, so an unseen
+# rewrite is unacceptable UX.  This gate converts agent-initiated (Hermes-tool-
+# spawned) writes into a y/N prompt in an interactive TTY, or a hard refusal in
+# the non-interactive context the agent actually runs in.  Human CLI writes
+# carry no Hermes-agent marker and are never gated.
+
+# Marker hand-inscribed by the agent command-execution layer (see
+# tools/environments/local.py).  A human's own ``hermes config set`` in their
+# login shell is never spawned by Hermes and never carries this marker.
+_HERMES_AGENT_INITIATED_ENV = "HERMES_AGENT_INITIATED"
+
+
+def _is_model_routing_key(key: str) -> bool:
+    """Return True when ``key`` routes model selection (billing-relevant).
+
+    Covers the model-routing keys an agent-initiated rewrite must not do
+    silently: ``model`` (shorthand for ``model.default``), ``model.*``,
+    ``delegation.model/provider/base_url``, and
+    ``auxiliary.<name>.model/provider/base_url``.
+    """
+    if not key:
+        return False
+    k = key.strip().lower()
+    if k == "model":
+        return True
+    if k.startswith("model."):
+        return True
+    if k in ("delegation.model", "delegation.provider", "delegation.base_url"):
+        return True
+    if k.startswith("auxiliary."):
+        parts = k.split(".")
+        return len(parts) == 3 and parts[2] in ("model", "provider", "base_url")
+    return False
+
+
+def _is_agent_initiated_write() -> bool:
+    """True when this process was spawned by a Hermes agent tool.
+
+    The agent command-execution layer stamps every subprocess it launches with
+    ``HERMES_AGENT_INITIATED=1``.  A human's direct ``hermes config set`` carries
+    no such marker.  Fail-open: any detection error returns False so a human
+    write is never mis-gated.
+    """
+    try:
+        marker = os.environ.get(_HERMES_AGENT_INITIATED_ENV, "").strip().lower()
+        return marker in ("1", "true", "yes")
+    except Exception:
+        return False
+
+
+def _model_config_confirm_enabled() -> bool:
+    """Read ``approvals.model_config_confirm`` (default True), fail-open to True."""
+    try:
+        cfg = load_config()
+        approvals = cfg.get("approvals") if isinstance(cfg, dict) else None
+        if isinstance(approvals, dict):
+            return bool(approvals.get("model_config_confirm", True))
+        return True
+    except Exception:
+        return True
+
+
+def _is_interactive_tty() -> bool:
+    """True when stdin is a real TTY.  Any detection error is treated as
+    non-interactive (the safe/refusing branch) so a gate can never hang."""
+    try:
+        return bool(sys.stdin.isatty())
+    except Exception:
+        return False
+
+
+def _lookup_old_value(key: str):
+    """Best-effort read of the current value for ``key`` (None-safe)."""
+    try:
+        found = _get_nested(load_config(), key)
+        return None if found is _MISSING else found
+    except Exception:
+        return None
+
+
+def _emit_model_change_notice(key: str, old_value, new_value) -> None:
+    """Surface the *fact* of a model-routing change, even when approved."""
+    old_str = "None" if old_value is None else old_value
+    print(f"⚠ {key} changed: {old_str} → {new_value} (model-routing config)")
+
+
+def _emit_model_routing_refusal(key: str, old_value, new_value) -> None:
+    """Non-interactive refusal: no hang, a loud instruction to surface it."""
+    old_str = "None" if old_value is None else old_value
+    print(
+        f"✗ Refusing agent-initiated change to model-routing key {key!r} "
+        f"({old_str} → {new_value}).\n"
+        "  Model-routing config determines what you pay per token for every "
+        "subagent, so an agent must not rewrite it silently.\n"
+        "  Surface this proposed change to the user and retry only on their "
+        "explicit direction with:\n"
+        f"      hermes config set {key} {new_value} --confirm-model-change",
+        file=sys.stderr,
+    )
+
+
+def _prompt_model_config_confirm(key: str, new_value) -> bool:
+    """Interactive y/N prompt (destructive-slash style)."""
+    prompt = f"⚠ Agent-initiated model-routing change: set {key} = {new_value!r}? [y/N] "
+    try:
+        answer = input(prompt)
+    except EOFError:
+        return False
+    return answer.strip().lower() in ("y", "yes")
+
+
+def _maybe_gate_model_routing_write(
+    key: str,
+    value,
+    *,
+    confirm_model_change: bool,
+    old_value,
+) -> bool:
+    """Gate an agent-initiated write to a model-routing key.
+
+    Returns True to allow the write, False to refuse it (message already
+    emitted).  Human/programmatic writes (no agent marker) pass through
+    untouched.  ``confirm_model_change`` (the CLI's ``--confirm-model-change``)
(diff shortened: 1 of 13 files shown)


alt-glitch:

Related: #97673 addresses the same model-routing confirmation goal at the approval layer; this PR uses an agent-origin marker and refusal path.

andrexibiza (commented):
Reviewed exact head `d4f62e806b92c0ee87d8df4c6dec1fa1e5e84bb6`, which is directly based on current `main@4209d371aa1bb8840ce8447555bdd863a1a96c38`. The core goal is right, and there is careful work here worth keeping: the pre-write refusal, cross-backend coverage, section-write coverage, and nested secret redaction are all useful hardening. I do not think this exact object is safe to land yet, though, because the authority proof is currently owned by the same command it is supposed to constrain.

### P1 — the agent/human provenance bit is ambient, caller-mutable state

`tools/environments/base.py::stamp_agent_initiated()` writes an ordinary `HERMES_AGENT_INITIATED=1` environment variable. The local/Docker paths propagate that variable, and the SSH/Modal/Daytona/Vercel paths prepend `export HERMES_AGENT_INITIATED=1; ...`. At the mutation boundary, `hermes_cli/config.py::_is_agent_initiated_write()` trusts that environment value to decide whether the write is an agent write.

But the agent owns the shell command that executes after the stamp. On this head, either of these command shapes changes the exact evidence the gate trusts before `hermes config set` consumes it:

```sh
HERMES_AGENT_INITIATED=0 hermes config set delegation.model <value>
env -u HERMES_AGENT_INITIATED hermes config set delegation.model <value>
```

The remote prefix has the same property: `export ...=1; <agent-controlled command>` establishes no stronger boundary because `<agent-controlled command>` can overwrite/unset it before spawning Hermes. `_is_agent_initiated_write()` then classifies the process as human/programmatic and `_maybe_gate_model_routing_write()` allows it.

The `--confirm-model-change` escape hatch is the same ownership problem in a second form: when the marker *is* present, `_maybe_gate_model_routing_write()` accepts the command's own boolean flag as sufficient proof of “user-directed” and immediately allows the write. The issue contract requires a distinction between an agent choosing a route and a user explicitly directing the exact change; that distinction cannot be authenticated by a flag controlled by the agent whose action is being gated.

This is the other side of the shape: the write sink is checking provenance, but the provenance has already crossed into ambient state controlled by the subject. Required fix: keep the agent/user principal and any user-direction permit at a boundary the child command cannot rewrite — e.g. the terminal/approval owner before process execution, or a scoped turn/session authority object validated there — and make the CLI flag at most a request that is checked against that stronger proof, not the proof itself. Please add a regression where an agent-origin command explicitly unsets/overrides the marker and still cannot mutate model routing, plus a regression proving an asserted bypass without actual approval is rejected pre-effect.

### P1 — the guarded principal can disable its own guard

There is a second complete bypass in the current policy surface. `_is_model_routing_key()` deliberately does **not** protect `approvals.model_config_confirm`, so an agent-marked process may run:

1. `hermes config set approvals.model_config_confirm false`
2. `hermes config set delegation.model <anything>`

and the second write is allowed. This is not just an inferred edge: `TestGateDisabled.test_gate_disabled_allows_agent_write` encodes exactly this sequence — it sets `HERMES_AGENT_INITIATED=1`, writes the opt-out to `false`, then verifies the agent can persist `delegation.model`.

A user-owned “Always Approve”/opt-out is reasonable; a subject-owned kill switch for the subject's own guard is not. The opt-out mutation itself needs to be owned by a stronger user/approval surface (or be protected by the same independent authority check), with a regression that an agent cannot self-disable the policy and then write routing state.

### Interlock / merge-order

There is already an independent competing implementation for the same closure: #97673 (`fangliquanflq`) also `Closes #97652`, but puts model-routing detection in `tools/approval.py` / `tools/terminal_tool.py` before command execution and routes it through the existing approval transports. #98127 and #97673 also directly collide in `hermes_cli/config_defaults.py` and `hermes_cli/subcommands/config.py`. These should not both independently land as owners of the same confirmation policy and both auto-close the issue. Please establish one explicit composition/supersession path; if #98127 absorbs or supersedes any of #97673's approach, preserve `fangliquanflq`'s independent contribution in that record. The strongest pieces look composable: the approval-layer ownership from #97673 and the backend/redaction/section-write regression work here.

#98128 is complementary rather than duplicate: per-dispatch model/provider overrides reduce the legitimate need for persistent `delegation.*` mutation but do not replace a consent boundary for persistent writes. #98130 is also complementary (validity of a model slug is not authority to persist it), but it changes `hermes_cli/config.py` on the same current-main base, so it needs an explicit merge order/recomposition with this PR rather than two independent edits to the same write path.

### Exact-head verification state

The PR is currently based exactly on current main, so there is no stale-base ambiguity. On exact head `d4f62e8`, Docker run `33272918339` and Nix run `33272918330` are green, but CI run `33272918792` is red: `Python tests / Run tests` failed, and the aggregate `All required checks pass` is therefore failed. That is an independent landing blocker until the repaired exact head has a fully green required-check set.

This is close to a useful policy boundary, and the review hardening already improved it materially. The remaining work is not broader feature scope; it is moving the proof of *who authorized the write* out of the mutable child process and making the guard's own policy state non-self-modifiable.

DavidMetcalfe:
@fangliquanflq @alt-glitch — Confirming the overlap is real: #97673 and this PR both gate agent-initiated writes to the same model-routing keys, both survive `approvals.mode: off`, and both add `approvals.model_config_confirm` (default true). #97673 was opened first (August 29, early UTC) — genuinely independent parallel work on the same issue, both closing #97652.

I ran #97673's test suite: 9 tests pass on its head, and disabling its detector fails exactly the 6 gate-path tests (restored green after).

The two PRs enforce at different layers:

- #97673 detects config-writing commands at the approval layer (parsing the agent's terminal commands, including compound and obfuscation-wrapped invocations) and routes them through the approval transports — an interactive confirm dialog on CLI/gateway surfaces, `--yes` for user direction, Always-Approve for opt-out persistence, and headless contexts fail closed (the write is refused when no approval surface exists).
- #98127 marks origin at spawn time (`HERMES_AGENT_INITIATED` stamped on every Hermes-spawned child env, covering all execution backends) and enforces inside `set_config_value` — agent-marked writes are refused outright (never prompting, since an agent process must not block on stdin), the user runs the change themselves, and the gate catches writes regardless of how they reach the config API (a `--force` section write or a shimmed subprocess both hit it).

So the headless trade-off is: #97673 refuses when no approval surface can render; #98127 always leaves the user-run path open. Both are defensible defaults — a maintainer call.

Factual note on combination: the two mechanisms coexist without conflict — origin-marker enforcement in the config API sets a refusal floor on every path, and approval-layer prompting adds an interactive dialog where a surface exists. Whichever direction maintainers pick, the label and layering decisions are theirs.

PR #80479, opened 2026-08-06
fix(delegation): resolve child fallback chains through the full pin x config matrix (#80450)

ayushnangia:
## Summary

Completes the delegation fallback-chain matrix from the #80450 cross-PR map: one pure function, `_resolve_child_fallback_chain`, implementing all six cells of the pin × `delegation.fallback_providers` decision table — composing the pin semantics of #80465 (@teknium1) with the config-chain semantics of #80438 (@wz-heng) / #80421 (@andrexibiza) for #65038 (@mlahatte), and settling the composition cell that all three leave undefined.

## Root cause (the class)

Every open PR on this seam answers "which fallback chain does a non-head agent get?" for one input combination, editing the same `parent_fallback` hunk. The combinations interact: #80465 nulls the chain on pin, #80438/#80421 install a configured chain — land them in either order and pinned-child semantics differ silently. The matrix needs to be decided once, in one place.

## Changes

`tools/delegate_tool.py` — `_resolve_child_fallback_chain(parent_agent, delegation_cfg, pinned)`:

| pin | `delegation.fallback_providers` | result |
|---|---|---|
| yes | declared, non-empty | declared chain — delegation-scoped recovery honors both explicit intents; the silent drag in #80450 is specifically the *parent's* chain substituting a pin |
| yes | declared, empty `[]` | none — explicit disable |
| yes | absent | none — pin fails loudly (exactly #80465's semantics) |
| no | declared, non-empty | declared chain (exactly #80438/#80421's semantics, #65038) |
| no | declared, empty `[]` | none — explicit disable |
| no | absent | inherit parent chain (historical behavior, byte-identical) |

Malformed declared values log and fall back **pin-aware** (none when pinned, parent chain otherwise) — extending #80421's log-and-inherit contract so the config-error path can't reintroduce the drag. Entries normalize through the canonical `hermes_cli.fallback_config.get_fallback_chain` (ordering/dedupe identical to top-level chains). The two pre-existing inheritance tests are made hermetic against the real user config, as #80438 also did.

## Validation

- `tests/tools/test_delegate_fallback_matrix.py` (14 tests): all six matrix cells, empty-parent edge, normalizer dedupe, pin-aware malformed fallback (both arms), non-dict config, plus four end-to-end wiring tests through `_build_child_agent` including pin+declared. Red on main: the suite fails collection (resolver doesn't exist); the wiring behaviors are unreachable.
[...]

Issue #80450, opened 2026-08-06
delegate_task: pinned delegation.provider/model silently replaced at runtime by inherited parent fallback chain


teknium1:
## Summary

Found during the pinned-model fallback audit item on tracker #79686. Spawn-time preflight for a pinned `delegation.provider`/`delegation.model` already fails loudly (good). But the pin is NOT honored after spawn: the child unconditionally inherits the parent's fallback chain, so a mid-run failure on the pinned provider silently reroutes the child onto the parent's fallback models.

## What already works (no gap at preflight)

`tools/delegate_tool.py`:
- `:3594-3604` — unresolvable pinned provider → `ValueError("Cannot resolve delegation provider '...'")`
- `:3607-3611` — resolved but no API key → `ValueError(...run 'hermes auth'...)`
- `:2879-2882` — caller returns `tool_error(...)`; spawn refused loudly.

## The gap (runtime substitution)

- `tools/delegate_tool.py:1469` — `parent_fallback = getattr(parent_agent, "_fallback_chain", None)` is inherited **unconditionally**, even when `override_provider` is set.
- `:1524` — passed to the child as `fallback_model=parent_fallback`.
- At runtime, `agent/chat_completion_helpers.py` `try_activate_fallback` (~L1730, swap at ~L1924-1927) replaces the child's `model/provider/base_url` with parent-chain entries on auth/429/5xx.
- The child runs `quiet_mode=True` (`:1527`) and the parent sees only the final summary — so results appear to come from the pinned model but didn't. No surfaced signal anywhere.

The asymmetry is right next door: `:1489-1495` already clears the parent's OpenRouter provider-preference filters when `override_provider` is set, precisely so parent routing can't silently drag the child back onto the parent's provider. The fallback chain needs the same treatment.

Secondary instance of the same class: `:1409-1419` — a pinned `delegation.command` (ACP transport) missing from PATH produces only `logger.warning` and silently falls back to the default transport.

## Proposed fix

Predictability > liveness for explicit pins:

- When `override_provider` (or an explicit `delegation.model` pin) is set, pass `fallback_model=None` — or a delegation-scoped fallback list if one is ever added — so the child fails loudly instead of silently substituting.
- Upgrade the ACP `delegation.command`-missing case from warning+fallback to a spawn failure with a clear message.

Related: tracker #79686 (audit item, "No-silent-fallback model pinning in delegate_task").

ayushnangia:
[...]

Issue #67347, opened 2026-07-19
[Feature]: Guided picker for Subagent Model + Provider in Advanced Settings (Desktop + Dashboard)


DavidMetcalfe:
# Guided picker for Subagent Model + Provider in Settings

## Problem

In **Settings → Advanced** (Hermes Desktop) and the **Config** page (Hermes Dashboard), the two fields that control subagent model behavior — `delegation.model` and `delegation.provider` — are exposed as bare free-text inputs. Users have no way to know what valid values look like:

- Which provider slugs does the gateway accept? (`openrouter`, `anthropic`, `custom`, `nous`, …)
- What's the model id format? (`anthropic/claude-sonnet-4-6`, `claude-sonnet-4-6`, or something else?)
- Is blank a valid value, and what does it do?

A typo silently does nothing because nothing validates, and the only signal of a misconfigured subagent is downstream (subagent uses the wrong model, or falls back to the parent). For users who want a *different* model for subagents (the common case — a cheaper or faster model than the main agent), this is a trial-and-error configuration surface.

The relevant config defaults make this concrete:

```yaml
# hermes_cli/config.py:2297-2298
delegation:
  model:    ""    # (empty = inherit parent model)
  provider: ""    # (empty = inherit parent provider + credentials)
```

…and the resolver at `tools/delegate_tool.py:3056-3170` confirms: when both fields are blank, `_resolve_delegation_credentials` returns `None` values and the child inherits everything from the parent. The fresh-install default is "inherit" — subagents are not disabled, but the UI gives no hint of that.

## Proposed Solution

Replace the two bare text inputs in both Hermes Desktop and Hermes Dashboard with a **paired provider + model picker**, mirroring the existing `FallbackModelsField` precedent in `apps/desktop/src/app/settings/fallback-models-field.tsx`:

- **Provider dropdown** sourced from `getGlobalModelOptions()` (the same data source the chat model picker, fallback-models field, and MoA presets already use). The provider entry shape (`ModelOptionProvider = {slug, name, models?, authenticated, is_user_defined?, …}`) is exactly what's needed.
- **Model dropdown** filtered by the selected provider, with free-text fallback so power users can type arbitrary model ids not in any catalog.
- **"Inherit from main agent"** surfaced as a first-class dropdown option that writes `""` to both keys (the documented default behavior).
[...]

Issue #81267, opened 2026-08-07
Cron + background delegate_task: shared SessionDB use-after-close silently drops child transcripts; completions unroutable (session_key=<cli>); missing Discord send/edit-message tool


nosstraa:
**Version:** hermes-agent 0.18.2, commit `29b8cacfa` (`v2026.7.7.2-510-g29b8cacfa`) · Linux, Python 3.12, gateway long-running under systemd
**Related:** follow-up to #79468 (same deployment). This report bundles two defects with a shared trigger plus one small tool gap surfaced by the same investigation.

---

### Symptom

`run_agent: Session DB append_message failed: 'NoneType' object has no attribute 'execute'` — 144 occurrences over two days, all in cron-spawned **subagent** sessions. The subagent keeps running and its filesystem work completes, but everything after the parent's teardown is silently dropped from the session DB (the failure is downgraded to a WARNING).

### Root cause (one sentence)

`cron/scheduler.py:run_job()` closes the per-job `SessionDB` in its `finally` block while a **fire-and-forget background delegation subagent** — which was handed the *same* `SessionDB` object — is still running on a daemon thread; every subsequent flush in the child calls `execute` on the closed connection (`self._conn = None`).

### Code path

Where the error is swallowed — `run_agent.py` ~1892–1914, `_flush_messages_to_session_db`:

```python
                self._session_db.append_message(
                    session_id=self.session_id,
                    role=role,
                    content=content,
                    ...
                )
                msg[_DB_PERSISTED_MARKER] = True
            ...
        except Exception as e:
            logger.warning("Session DB append_message failed: %s", e)
```

The `except` wraps the whole write loop — one failure aborts the entire flush, downgraded to a warning, so the agent keeps running with zero persistence.

Where the `None` comes from — the **only** assignment of `_conn = None` outside `__init__` is `SessionDB.close()` (`hermes_state.py` ~1318–1331):

```python
    def close(self):
        ...
            if self._conn:
                try:
                    self._conn.execute("PRAGMA wal_checkpoint(TRUNCATE)")
                except Exception:
                    pass
                self._conn.close()
                self._conn = None
```

Who closes it — cron job teardown (`cron/scheduler.py` ~2620 creates one SessionDB per job, ~3076 hands it to the job's AIAgent, ~3336ff `finally:` calls `end_session(...)` then `_session_db.close()` at ~3354).

[...]

PR #98128, opened 2026-08-29
feat(delegation): per-dispatch model/provider override for delegate_task

### LINKED ISSUE #97653: [Feature]: Per-dispatch model/provider override for delegate_task subagents

DavidMetcalfe:
## What does this PR do?

Adds optional per-task `model` and `provider` parameters to `delegate_task`'s
`tasks[]` items, giving agents a legitimate per-dispatch routing lever so they
never need to rewrite persistent `delegation.model` config to get a different
model for one job.

## Motivation

`delegate_task` tasks accept only `goal`/`context`/`role`/`toolsets`. When the
configured delegation model fails (or a task needs a vision/cheaper/faster
model), an agent's only lever today is `hermes config set delegation.model ...`
— which silently reroutes *all* future subagents in every session. A real
incident on my install: a session hit `HTTP 400: not a valid model ID` on a
delegation batch, "repaired" the config by switching models, and the persistent
change went unnoticed for days. A per-dispatch override makes the legitimate
response ("re-dispatch on a working model") possible with zero persistent side
effects.

## Rebase note (2026-09-07)

This PR was filed against the pre-split monolithic `tools/delegate_tool.py`.
While it sat in review, main restructured the delegation surface:

- **Facade + siblings**: `delegate_tool.py` is now a facade over nine
  `delegate_tool_*.py` siblings; child construction moved into `_build_children()`.
- **#105347 / #80450** (salvage of #80479): `_resolve_child_fallback_chain()` landed
  a pin-aware fallback contract — pinned children never borrow the parent's
  fallback chain; unpinned children inherit it; explicit `[]` disables either way.

Before rebasing, I probed whether main had superseded the feature:
`_dispatch_delegate_task` and `_normalize_task_list` still pass task dicts
through unfiltered, no per-task `model`/`provider` resolution exists anywhere on
main, and the `tasks[]` schema has no override fields — so the feature is still
needed and **composes** with (not duplicates) the landed fallback work: the new
`pinned=bool(override_provider or override_base_url or model)` in
`_resolve_child_runtime()` already gives per-dispatch overrides the
#105347 semantics (a per-task model-only override counts as a pin, so an
overridden child never silently escalates through the parent's fallback chain).

A take-main rebase dropped the feature code entirely (the construction loop the
original diff patched no longer exists), so the two-pass resolution was
**deliberately re-ported** onto the new structure: pass-1 per-task resolution
lives in `delegate_task()`, and per-task creds thread into `_build_children()`
via a new `resolved_task_creds` parameter. The rework is one commit on current
main tip `2237be3559`; the original review history below is preserved. The
413-line regression test file is unchanged from the originally reviewed
version and passes 10/10.

## Design

- **Precedence**: per-task override > `delegation.model/provider` config >
  parent inheritance. Unset behaves exactly as today.
- **Two-pass resolution**: every task's override resolves through the existing
  `_resolve_delegation_credentials` → `resolve_runtime_provider` path *before*
  any child is constructed. An invalid override fails the whole call with a
  per-task-index error and constructs nothing — no orphaned children, no
  leaked SessionDB handles, no half-run batches. Pass 1 runs before transcript
  creation and the batch announcement, so a failed dispatch writes no live
  logs and announces nothing.
- **No persistence**: the override is dispatch-scoped; a guard test asserts
  `save_config` is never invoked on this path.
- **Concurrency unchanged**: overrides don't bypass `max_concurrent_children`.
- **Footprint**: two compact optional string params (AGENTS.md narrow-waist
  rule), each description ~2 lines.

## Changes Made

- `tools/delegate_tool.py` — `tasks[]` item schema gains `model`/`provider`;
  two-pass override resolution in `delegate_task()` (merging each task's
  override over the same `routing_cfg` the batch-level creds resolve from, so
  internal callers like `/review` keep their routing owner); per-task creds
  threading through `_build_children()`; broadened exception attribution
  (`logger.exception` before the per-task `tool_error` so unexpected resolution
  errors stay diagnosable).
- `tests/tools/test_delegate_model_override.py` — 10 tests: schema exposure,
  precedence, inheritance fallback, model-only override preserving config
  provider, invalid-override isolation (all-or-nothing,
  `MockAgent.assert_not_called`), non-ValueError attribution, mixed-batch
  independence, credential-pool graceful path for non-pool providers,
  `save_config` never called.

## How to Test

1. Batch with per-task `model`/`provider` overrides → each child runs on its
   resolved provider (visible in the batch report's per-task `Model:` line).
2. Batch with one invalid override → single `tool_error` naming the task
   index; no children constructed, no transcripts created.
3. Batch without overrides → byte-identical behavior to current main.

`scripts/run_tests.sh` over the 28 delegate-adjacent test files: 610 passed,
1 failed — `test_delegate.py::test_child_dedicated_db_follows_parents_db_path`,
which also fails on clean `origin/main` (verified on a detached worktree at
`2237be3559`); it exercises `_build_child_agent` directly and is unrelated to
this diff.

## Review

Three rounds of cross-vendor review (Gemini Flash High + GPT-OSS 120B Medium,
full diffs in byte-identical briefs).

- Rounds 1–2 (original diff): both reviewers ACCEPTABLE; GPT-OSS's remaining
  should-fix (log unexpected exceptions before flattening to tool_error)
  applied. Two round-1 GPT-OSS claims rejected after source verification
  (credential-pool handling already graceful; no persistence path exists),
  documented in the commit bodies.
- Round 3 (this rework, on the post-#105347 structure): both reviewers READY.
  GPT-OSS flagged the broad `except Exception` as SHOULD-FIX; declined with
  evidence — the breadth is load-bearing (`resolve_runtime_provider` raises
  `RuntimeError` for missing keys/providers, two tests pin the
  RuntimeError → clean-tool_error contract, and the log-first pattern was
  round-2's own ask). Flash's two NITs (dead `creds` param in
  `_build_children`; pass-1 placement before transcript creation) applied.

Closes #97653

Part of #67347 (the per-dispatch override is the deliberately-nonpersistent
complement to the user-saved Models-pane preferences that issue describes).

### DIFF
--- a/tests/tools/test_delegate_model_override.py
+++ b/tests/tools/test_delegate_model_override.py
@@ -1,0 +1,413 @@
+#!/usr/bin/env python3
+"""
+Per-dispatch model/provider override for delegate_task subagents.
+
+Issue #97653: each entry in a delegate_task `tasks` batch may carry an
+optional per-task `model` and/or `provider`. The override is resolved through
+the same runtime-provider path as delegation config (resolve_runtime_provider),
+applies ONLY to that dispatch's child(ren), and is NEVER persisted to config.
+Precedence: per-task override > delegation config > parent inheritance.
+
+Run with: python -m pytest tests/tools/test_delegate_model_override.py -v
+"""
+
+import json
+import threading
+import unittest
+from unittest.mock import MagicMock, patch
+
+from tools.delegate_tool import (
+    DELEGATE_TASK_SCHEMA,
+    _resolve_child_credential_pool,
+    delegate_task,
+)
+
+
+def _make_mock_parent(depth=0):
+    parent = MagicMock()
+    parent.base_url = "https://openrouter.ai/api/v1"
+    parent.api_key = "***"
+    parent.provider = "openrouter"
+    parent.api_mode = "chat_completions"
+    parent.model = "anthropic/claude-sonnet-4"
+    parent.platform = "cli"
+    parent.providers_allowed = None
+    parent.providers_ignored = None
+    parent.providers_order = None
+    parent.provider_sort = None
+    parent._session_db = None
+    parent._delegate_depth = depth
+    parent._active_children = []
+    parent._active_children_lock = threading.Lock()
+    parent._print_fn = None
+    parent.tool_progress_callback = None
+    parent.thinking_callback = None
+    return parent
+
+
+def _fake_resolve(requested=None, target_model=None):
+    """Synthetic runtime-provider bundle keyed on the requested provider name."""
+    provider = requested or "openrouter"
+    return {
+        "provider": provider,
+        "model": target_model or "resolved-model",
+        "base_url": f"https://{provider}.example/v1",
+        "api_key": f"{provider}-key",
+        "api_mode": "chat_completions",
+    }
+
+
+def _completed(idx):
+    return {
+        "task_index": idx,
+        "status": "completed",
+        "summary": "ok",
+        "api_calls": 1,
+        "duration_seconds": 1.0,
+        "_child_role": None,
+    }
+
+
+class TestPerDispatchOverrideSchema(unittest.TestCase):
+    def test_tasks_items_accept_model_and_provider_overrides(self):
+        """The tasks[] item schema advertises the optional override params."""
+        task_props = (
+            DELEGATE_TASK_SCHEMA["parameters"]["properties"]["tasks"]["items"]["properties"]
+        )
+        self.assertIn("model", task_props)
+        self.assertEqual(task_props["model"]["type"], "string")
+        self.assertIn("provider", task_props)
+        self.assertEqual(task_props["provider"]["type"], "string")
+
+
+class TestPerDispatchOverrideResolution(unittest.TestCase):
+    """Per-task override resolves through the runtime-provider path and wins
+    over the delegation config (and over parent inheritance)."""
+
+    @patch("tools.delegate_tool._load_config")
+    @patch("hermes_cli.runtime_provider.resolve_runtime_provider")
+    def test_override_wins_over_config(self, mock_resolve, mock_cfg):
+        mock_resolve.side_effect = _fake_resolve
+        # Config pins openrouter/config-model as the routing baseline.
+        mock_cfg.return_value = {
+            "max_iterations": 45,
+            "model": "config-model",
+            "provider": "openrouter",
+        }
+        parent = _make_mock_parent()
+
+        with (
+            patch("run_agent.AIAgent") as MockAgent,
+            patch("tools.delegate_tool._run_single_child") as mock_run,
+        ):
+            mock_child = MagicMock()
+            MockAgent.return_value = mock_child
+            mock_run.return_value = _completed(0)
+            delegate_task(
+                tasks=[{"goal": "Resolve the override flags", "model": "override-model",
+                        "provider": "override-prov"}],
+                parent_agent=parent,
+            )
+
+        _, kwargs = MockAgent.call_args
+        # The override beat config (and parent provider=openrouter).
+        self.assertEqual(kwargs["model"], "override-model")
+        self.assertEqual(kwargs["provider"], "override-prov")
+        self.assertEqual(kwargs["base_url"], "https://override-prov.example/v1")
+        self.assertEqual(kwargs["api_key"], "override-prov-key")
+
+    @patch("tools.delegate_tool._load_config")
+    @patch("hermes_cli.runtime_provider.resolve_runtime_provider")
+    def test_empty_or_missing_override_inherits_config(self, mock_resolve, mock_cfg):
+        mock_resolve.side_effect = _fake_resolve
+        # Config pins a DIFFERENT provider than the parent so inherit-config
+        # (not inherit-parent) is observable.
+        mock_cfg.return_value = {
+            "max_iterations": 45,
+            "model": "config-model",
+            "provider": "nous",
+        }
+        parent = _make_mock_parent()  # parent provider=openrouter, model=anthropic/...
+
+        with (
+            patch("run_agent.AIAgent") as MockAgent,
+            patch("tools.delegate_tool._run_single_child") as mock_run,
+        ):
+            mock_child = MagicMock()
+            MockAgent.return_value = mock_child
+            mock_run.return_value = _completed(0)
+            delegate_task(
+                tasks=[{"goal": "Resolve inherits when override is empty", "model": "", "provider": ""}],
+                parent_agent=parent,
+            )
+
+        _, kwargs = MockAgent.call_args
+        self.assertEqual(kwargs["model"], "config-model")
+        self.assertEqual(kwargs["provider"], "nous")
+
+    @patch("tools.delegate_tool._load_config")
+    @patch("hermes_cli.runtime_provider.resolve_runtime_provider")
+    def test_model_only_override_preserves_config_provider(self, mock_resolve, mock_cfg):
+        mock_resolve.side_effect = _fake_resolve
+        mock_cfg.return_value = {
+            "max_iterations": 45,
+            "model": "config-model",
+            "provider": "openrouter",
+        }
+        parent = _make_mock_parent()
+
+        with (
+            patch("run_agent.AIAgent") as MockAgent,
+            patch("tools.delegate_tool._run_single_child") as mock_run,
+        ):
+            mock_child = MagicMock()
+            MockAgent.return_value = mock_child
+            mock_run.return_value = _completed(0)
+            delegate_task(
+                tasks=[{"goal": "Override only the model", "model": "override-model"}],
+                parent_agent=parent,
+            )
+
+        _, kwargs = MockAgent.call_args
+        self.assertEqual(kwargs["model"], "override-model")
+        self.assertEqual(kwargs["provider"], "openrouter")
+
+    @patch("tools.delegate_tool._load_config")
+    @patch("hermes_cli.runtime_provider.resolve_runtime_provider")
+    def test_invalid_override_fails_that_dispatch_without_config_write(self, mock_resolve, mock_cfg):
+        mock_cfg.return_value = {
+            "max_iterations": 45,
+            "model": "config-model",
+            "provider": "openrouter",
+        }
+
+        def _resolve(**kwargs):
+            if kwargs.get("requested") == "bogus-prov":
+                raise RuntimeError("BOGUS_PROVIDER_API_KEY not set")
+            return _fake_resolve(**kwargs)
+
+        mock_resolve.side_effect = _resolve
+        parent = _make_mock_parent()
+
+        with patch("run_agent.AIAgent") as MockAgent:
+            result = json.loads(
+                delegate_task(
+                    tasks=[{"goal": "Resolve with a bad provider", "provider": "bogus-prov"}],
+                    parent_agent=parent,
+                )
+            )
+            # The invalid override fails the dispatch loudly — no child is
+            # spawned, so nothing runs on the wrong provider.
+            MockAgent.assert_not_called()
+
+        self.assertIn("error", result)
+        self.assertIn("Task 0", result["error"])
+        self.assertIn("bogus-prov", result["error"])
+        # No config write happens: the override code path only READS config
+        # (_load_config / load_config_readonly); there is no writer call site.
+        self.assertEqual(mock_cfg.return_value["provider"], "openrouter")
+
+    @patch("tools.delegate_tool._load_config")
+    @patch("hermes_cli.runtime_provider.resolve_runtime_provider")
+    def test_mixed_overrides_resolve_each_child_independently(self, mock_resolve, mock_cfg):
+        mock_resolve.side_effect = _fake_resolve
+        mock_cfg.return_value = {
+            "max_iterations": 45,
+            "model": "config-model",
+            "provider": "openrouter",
+        }
+        parent = _make_mock_parent()
+
+        with (
+            patch("run_agent.AIAgent") as MockAgent,
+            patch("tools.delegate_tool._run_single_child") as mock_run,
+        ):
+            mock_child = MagicMock()
+            MockAgent.return_value = mock_child
+            mock_run.side_effect = [_completed(0), _completed(1)]
+            result = json.loads(
+                delegate_task(
+                    tasks=[
+                        {"goal": "First task with a pinned provider",
+                         "model": "m0", "provider": "p0"},
+                        {"goal": "Second task inherits the delegation config"},
+                    ],
+                    parent_agent=parent,
+                )
+            )
+
+        self.assertNotIn("error", result)
+        self.assertEqual(len(MockAgent.call_args_list), 2)
+        # Children are constructed in task order; each resolves independently.
+        _, kw0 = MockAgent.call_args_list[0]
+        self.assertEqual(kw0["model"], "m0")
+        self.assertEqual(kw0["provider"], "p0")
+        _, kw1 = MockAgent.call_args_list[1]
+        self.assertEqual(kw1["model"], "config-model")
+        self.assertEqual(kw1["provider"], "openrouter")
+
+
+class TestPerDispatchOverrideFailureIsolation(unittest.TestCase):
+    """Errors during per-task override resolution must fail the dispatch
+    cleanly and never orphan a prior constructed child or write config."""
+
+    @patch("tools.delegate_tool._load_config")
+    @patch("tools.delegate_tool._resolve_delegation_credentials")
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related: #77953 already implements the broader per-call and per-task model/provider override surface; this PR focuses the task-level pre-resolution path.

DavidMetcalfe:
@alt-glitch — Clarifying the relationship: #77953 is still open and unmerged, so the per-dispatch override surface isn't on main yet.

For anyone comparing the two PRs: both add per-task `model`/`provider` overrides on `delegate_task` with the same precedence ordering — @apoapostolov's #77953 got there first (August 3). They differ on failure handling, which is the part a rebase can't reconcile:

- #77953 resolves per-task credentials inside the dispatch loop and returns `tool_error(...)` mid-loop, so an invalid override on task N leaves tasks 0..N-1 already constructed. Constructing a child opens a dedicated SessionDB handle (#81267) and registers it on `parent._active_children`; a constructed-but-never-run child leaks both. The same concern was raised as an observation on #77953's own review thread and remains unresolved there.
- #98128 resolves every override in a first pass before any child is constructed, so dispatch is all-or-nothing: one invalid override fails the whole call with a per-task-index error and nothing is constructed or leaked. `test_mixed_batch_invalid_override_never_constructs_prior_child` pins the isolation, and a guard test asserts `save_config` is never called.

Happy to coordinate on a single PR if the maintainers prefer #77953's shape — the mid-batch orphan construction is the one part that needs addressing either way.

Joeywrz:
Thanks for the two-pass preflight and the focused scope here. I reviewed head `a2ab9fb5c2f80c1d69536dbd6e386d20a157394a` with Hermes assistance and prepared a small additive regression/fix proposal **on top of this PR**, rather than opening another competing implementation.

### Reproduced: a task provider override can retain the configured direct endpoint

At the per-task merge, `dict(routing_cfg)` retains `base_url`, `api_key`, `api_mode`, and `request_overrides`. Replacing only `provider` therefore still enters the resolver's direct-endpoint branch for non-native-SDK providers. In a real temporary configuration, switching from `custom:route-a` to `custom:route-b` retained the direct endpoint instead of resolving route B; a missing custom-provider override also failed to reject the batch.

The proposed six-line production addition drops those route-specific fields only when the explicit task provider differs from the configured provider token. Model-only, same-provider, and unmodified sibling tasks preserve their direct configuration. It does not change global direct-endpoint precedence or persisted configuration. Provider-alias equivalence is not established by this patch; the comparison deliberately matches configured tokens after case/whitespace normalization.

### Regression evidence

The two added test functions exercise the real registry dispatch, config loader, provider resolver, and child-construction path using a temporary `HOME`/`HERMES_HOME`, synthetic custom providers and keys, and `.invalid` endpoints. Only `AIAgent` is replaced; socket connections are forbidden. They verify route coherence, unchanged sibling/config state, and failure before any child for an invalid later task.

- Unmodified PR head plus new tests: **3 failed, 4 passed** (the two route-switch parameter cases and invalid-provider case fail).
- With proposed fix: **7 new cases passed**; complete override test file **17 passed**, independently rerun by the parent reviewer.
- Affected-area run: **578 passed, 1 failed** across 45 files. The remaining `test_child_dedicated_db_follows_parents_db_path` failure is the macOS `/private/tmp` versus `/tmp` string comparison and reproduces on the unchanged PR head. It is not hidden or fixed in this proposal.
- Ruff, compile check, Windows-footgun check, compatibility-pointer check, and `git diff --check` passed.
- A separate Hermes reviewer inspected the complete additive diff and recorded no blocking findings. This is AI-assisted review, not maintainer approval.

**Limits:** no full-suite run, no live provider requests, no Linux/Windows runtime verification, and no hosted CI for this additive patch. Existing PR CI does not certify the patch below.

Focused command (using the repository's hermetic wrapper):
```sh
scripts/run_tests.sh tests/tools/test_delegate_model_override.py -q
```

<details>
<summary>Additive patch against the reviewed PR head (two files)</summary>

```diff
diff --git a/tests/tools/test_delegate_model_override.py b/tests/tools/test_delegate_model_override.py
index e88af0697..02be960f2 100644
--- a/tests/tools/test_delegate_model_override.py
+++ b/tests/tools/test_delegate_model_override.py
@@ -11,11 +11,16 @@ Precedence: per-task override > delegation config > parent inheritance.
 Run with: python -m pytest tests/tools/test_delegate_model_override.py -v
 """
 
+import copy
 import json
+import socket
 import threading
 import unittest
 from unittest.mock import MagicMock, patch
 
+import pytest
+import yaml
+
 from tools.delegate_tool import (
     DELEGATE_TASK_SCHEMA,
     _resolve_child_credential_pool,
@@ -409,5 +414,111 @@ class TestPerDispatchOverrideFailureIsolation(unittest.TestCase):
         mock_save.assert_not_called()
 
 
+@pytest.fixture
+def direct_route(tmp_path, monkeypatch):
+    """Real config and provider resolution; only the child runtime is replaced."""
+    from hermes_cli.config import load_config_readonly
+    from tools.registry import registry
+
+    monkeypatch.setenv("HOME", str(tmp_path))
+    monkeypatch.setenv("HERMES_HOME", str(tmp_path))
+    monkeypatch.delenv("HERMES_IGNORE_USER_CONFIG", raising=False)
+    config = {
+        "delegation": {
+            "model": "configured-model",
+            "provider": "custom:route-a",
+            "base_url": "https://direct.invalid/v1",
+            "api_key": "direct-fixture-key",
+            "api_mode": "anthropic_messages",
+            "request_overrides": {"extra_body": {"route": "direct"}},
+            "max_spawn_depth": 2,
+        },
+        "providers": {
+            name: {
+                "base_url": f"https://{name}.invalid/v1",
+                "api_key": f"{name}-fixture-key",
+                "extra_body": {"route": name},
+            }
+            for name in ("route-a", "route-b")
+        },
+    }
+    path = tmp_path / "config.yaml"
+    path.write_text(yaml.safe_dump(config), encoding="utf-8")
+    before = path.read_bytes()
+    cached = load_config_readonly()["delegation"]
+    original = copy.deepcopy(cached)
+    parent = _make_mock_parent(depth=1)  # Registry waits for nested delegates.
+    parent._credential_pool = None
+    parent._current_task_id = None
+    parent.session_id = "fixture-parent"
+
+    def no_network(*args, **kwargs):
+        raise AssertionError("Provider calls are forbidden in routing tests")
+
+    monkeypatch.setattr(socket.socket, "connect", no_network)
+    with patch("run_agent.AIAgent") as agent:
+        def child(**kwargs):
+            instance = MagicMock()
+            instance.model = kwargs["model"]
+            instance.session_id = "fixture-child"
+            instance._credential_pool = None
+            instance.run_conversation.return_value = {
+                "final_response": "ok", "completed": True, "api_calls": 0,
+            }
+            return instance
+
+        agent.side_effect = child
+        yield registry, parent, agent, cached, tmp_path
+    assert path.read_bytes() == before
+    assert cached == original
+
+
+@pytest.mark.parametrize("provider", ["custom:route-b", "custom:route-a", ""])
+@pytest.mark.parametrize("model", ["task-model", None])
+def test_real_resolver_keeps_each_task_route_coherent(direct_route, provider, model):
+    registry, parent, agent, cfg, _ = direct_route
+    result = json.loads(registry.dispatch(
+        "delegate_task",
+        {"tasks": [
+            {"goal": "Check the overridden route", "model": model, "provider": provider},
+            {"goal": "Check the inherited route"},
+        ]},
+        parent_agent=parent,
+    ))
+    assert "error" not in result, result
+    assert len(agent.call_args_list) == 2
+    overridden, inherited = [call.kwargs for call in agent.call_args_list]
+    assert overridden["model"] == (model or cfg["model"])
+    if provider == "custom:route-b":
+        assert overridden["provider"] == provider
+        assert overridden["base_url"] == "https://route-b.invalid/v1"
+        assert overridden["api_key"] == "route-b-fixture-key"
+        assert overridden["api_mode"] == "chat_completions"
+        assert overridden["request_overrides"] == {"extra_body": {"route": "route-b"}}
+    else:
+        for key in ("base_url", "api_key", "api_mode", "request_overrides"):
+            assert overridden[key] == cfg[key]
+    for key in ("model", "base_url", "api_key", "api_mode", "request_overrides"):
+        assert inherited[key] == cfg[key]
+
+
+def test_real_resolver_rejects_invalid_override_before_any_child(direct_route):
+    registry, parent, agent, _, home = direct_route
+    result = json.loads(registry.dispatch(
+        "delegate_task",
+        {"tasks": [
+            {"goal": "Resolve a valid provider", "provider": "custom:route-b"},
+            {"goal": "Reject an invalid provider", "provider": "custom:missing-fixture-provider"},
+        ]},
+        parent_agent=parent,
+    ))
+    assert "error" in result, result
+    assert "Task 1" in result["error"]
+    assert "missing-fixture-provider" in result["error"]
+    agent.assert_not_called()
+    assert parent._active_children == []
+    assert not list(home.glob("cache/delegation/live/*"))
+
+
 if __name__ == "__main__":
     unittest.main()
diff --git a/tools/delegate_tool.py b/tools/delegate_tool.py
index fc3468ad2..b724f2e75 100644
--- a/tools/delegate_tool.py
+++ b/tools/delegate_tool.py
@@ -501,6 +501,12 @@ def delegate_task(
             if task_model_override:
                 per_task_cfg["model"] = task_model_override
             if task_provider_override:
+                if task_provider_override.lower() != str(routing_cfg.get("provider") or "").strip().lower():
+                    # A provider switch must not reuse the previous route's endpoint,
+                    # credentials or request personality. Model-only and same-provider
+                    # overrides still honor an explicitly configured direct endpoint.
+                    for key in ("base_url", "api_key", "api_mode", "request_overrides"):
+                        per_task_cfg.pop(key, None)
                 per_task_cfg["provider"] = task_provider_override
             try:
                 resolved_task_creds.append(
```
</details>

Happy for this to be folded into your existing PR if this is the intended task-override contract. The routing feature and its original authorship remain yours; this contribution is limited to the regression and boundary fix above.

AlirezaM02:
Hello, and thanks for rebasing this onto the split delegation architecture.

I independently built a related implementation before finding this PR, so I wanted to coordinate before opening competing work:

https://github.com/AlirezaM02/hermes-agent/tree/feat/model-aware-delegation

I originally built it for token-usage management, then generalized it into a model-aware delegation design. The workflow differs slightly:

- task-level `model`
- operator-configured `delegation.model_categories` and `delegation.model_tags`
- task selectors `category` and `model_tag`
- precedence: `model` > known `category` > known `model_tag` > normal delegation route

A `{model, provider}` alias resolves as a fresh provider route, so the prior route's endpoint, credentials, API mode, ACP settings, fallback chain, and request settings do not carry into the new provider. This is relevant to the provider-route leakage regression described in the recent comment above.

The alias approach keeps provider selection operator-controlled rather than exposing an unrestricted task-level `provider` field to the model. It also includes strict selector/config validation, neutral metadata for heterogeneous batches, documentation, and targeted tests.

Would maintainers prefer that I:

1. contribute route-isolation tests and hardening as follow-up work on this PR, or
2. open the alias-based implementation as a separate PR for comparison?

I will avoid opening a duplicate PR until there is direction.

DavidMetcalfe:
@Joeywrz — the route-leak is confirmed and I've adopted your fix plus the two regression tests on this branch.

Mechanism: `per_task_cfg = dict(routing_cfg)` kept the batch config's `base_url`/`api_key`/`api_mode`/`request_overrides`, so a `provider` override to a non-native-SDK token still hit `_resolve_delegation_credentials`' direct-endpoint branch — which re-derives `provider="custom"` from the retained base_url and ignores the override. On a provider switch the child resolved the *previous* route's endpoint and credentials rather than the requested provider's, and an invalid override token never rejected the batch.

Evidence (run on the PR head via `scripts/run_tests.sh`):
- Mutation check (fix reverted): 3 failed / 14 passed — the two route-switch cases assert `'custom' == 'custom:route-b'`, and `test_real_resolver_rejects_invalid_override_before_any_child` stops erroring because the invalid override is silently dropped onto the retained direct endpoint.
- With the fix: 17 passed (the full `tests/tools/test_delegate_model_override.py`).
- Broader delegate-area run (7 delegate test files): 80 passed / 1 failed — `test_child_dedicated_db_follows_parents_db_path` fails on the macOS `/private/tmp` vs `/tmp` path comparison and reproduces on the unchanged PR head; unrelated to this change.

The change is committed with you credited in the commit message. Thanks for the targeted reproducer — these route-field semantics weren't pinned by the existing tests, which all mock the resolver seam.

DavidMetcalfe:
@AlirezaM02 — thanks for flagging the parallel work and the operator-controlled-alias design point.

On the current thread: @Joeywrz found and contributed a route-isolation fix (a task `provider` override could retain the batch config's direct endpoint/credentials and silently defeat the override) and it is now folded into this PR, so option 1's hardening part is covered.

The design distinction you're raising is real: this PR exposes a task-level `provider` directly, while your alias approach keeps provider selection operator-controlled via `delegation.model_categories`/`model_tags` and resolves a `{model, provider}` alias as a fresh route. That operator-controlled-vs-task-exposed tradeoff is the maintainers' call, not mine. Either shape is cooperative — if the maintainers prefer the alias approach, I'm glad to build the route-isolation hardening on top of it rather than keep this PR's exposed `provider` field. Your route-isolation tests and the alias design are both useful follow-up material either way.

I'm not going to lobby for a shape. This thread is a good place for the maintainers to pick a direction before anyone opens a second overlapping PR.

PR #98129, opened 2026-08-29
fix(delegation): honest subagent failure statuses + config-level model-rejection notice

### LINKED ISSUE #97654: [Bug]: Delegation model rejections (model_not_found) are invisible to the user — no config-level notice

### LINKED ISSUE #97655: [Bug]: Failed subagents are mislabeled status=completed with exit_reason=max_iterations

DavidMetcalfe:
## What does this PR do?

Two fixes to the delegation failure pipeline, closing issues #97655 and
#97654. Both were found the same way: a misconfigured (hyphenated)
`delegation.model` caused every subagent to die in <1s with
`HTTP 400: not a valid model ID` — and the parent-facing reports *hid* both
the failure and its cause.

## Fix 1 — honest statuses for failed children (#97655)

A child that died in 0.74s with `api_calls=1` was reported as
`status=completed`, `exit_reason=max_iterations`, "TRUNCATED: hit
max_iterations". All three labels false. Root cause (two mis-derivations in
`_run_single_child`):

1. Status keyed only on summary presence — the provider's error text *is* the
   child's `final_response`, so error → "completed". The structured
   `result['failed']`/`result['error']` fields (set by
   `agent/conversation_loop.py` on every failure path) were never consulted.
2. `exit_reason` defaulted to `max_iterations` whenever the child's
   `completed` was False — but every failure path also returns
   `completed=False`.

Fix: structured fields consulted first (`failed`/`error` → `status="failed"`,
`exit_reason="error"`); `max_iterations` now means *only* genuine budget
exhaustion; `interrupted` unchanged; genuine truncation (budget exhausted with
a real summary) still reports `completed` + `truncated` — pinned by a
regression test.

## Fix 2 — config-level rejection notice (#97654)

The batch renderer now detects "≥1 task failed with a model_not_found-class
error naming the configured delegation model" and prepends a single actionable
warning:

```
⚠ SUBAGENT MODEL REJECTED: the configured Subagent Model "<model>" was
  rejected by provider "<provider>" (not a valid model ID). ...
  Check Settings → Advanced → Subagent Model, or: hermes config get delegation.model
```

with a "no fallback chain configured, so no failover was attempted" line when
the parent's chain is empty (the default). Detection matches
`agent/error_classifier.py`'s `_MODEL_NOT_FOUND_PATTERNS` and requires the
configured model's name in the error text, so a stale task failing on a
*different* model isn't mis-attributed. Emitted once per batch, both
renderers (fan-out and single-dispatch).

## Changes Made

- `tools/delegate_tool.py` — `_run_single_child` status/exit_reason branch
  reorder + docstring enum (`status ∈ {completed, interrupted, failed}`,
  `exit_reason ∈ {completed, max_iterations, interrupted, error}`).
- `tools/process_registry.py` — config-level notice in both delegation
  renderers; renderer unchanged otherwise (it derives ✗/⚠ correctly from the
  now-honest fields).
- Tests: 8 new across `tests/tools/test_delegate.py` +
  `tests/tools/test_process_registry.py`, including edge cases (error without
  `failed` key; empty `error` string + summary → still completed; mixed
  batch; notice emitted exactly once; no notice for non-model errors).

## Review

Two rounds of cross-vendor review (Gemini 3.7 Flash High + GPT-OSS 120B
Medium). Round-2 verdicts: ACCEPTABLE from both reviewers on this delta
("ready to merge"); the full branch went through the same loop with every
round-1 finding either fixed or rejected-with-source-evidence (dispositions in
commit bodies). 301+ passed on the delegate/process_registry suites.

Closes #97655
Closes #97654

### DIFF
--- a/tests/tools/test_delegate.py
+++ b/tests/tools/test_delegate.py
@@ -676,6 +676,107 @@ class TestDelegateObservability(unittest.TestCase):
             self.assertEqual(result["results"][0]["status"], "failed")
 
 
+class TestDelegateFailedChildStatus(unittest.TestCase):
+    """Honest status / exit_reason for failed subagents (issue #97655).
+
+    A child that fails on its first API call (e.g. an HTTP 400 "not a valid
+    model ID") returns completed=False with failed=True + an error string as
+    its terminal final_response. It must be reported as status=failed with an
+    honest exit_reason — never status=completed + exit_reason=max_iterations
+    (which mislabels provider rejections as iteration-budget exhaustion and
+    would render the false "TRUNCATED" banner).
+    """
+
+    def _delegate_single(self, child_result):
+        """Dispatch a single task whose mock child returns `child_result`,
+        returning the parsed child result entry dict."""
+        parent = _make_mock_parent(depth=0)
+        with patch("run_agent.AIAgent") as MockAgent:
+            mock_child = MagicMock()
+            mock_child.model = "claude-sonnet-4-6"
+            mock_child.session_prompt_tokens = 0
+            mock_child.session_completion_tokens = 0
+            mock_child.run_conversation.return_value = child_result
+            MockAgent.return_value = mock_child
+            result = json.loads(
+                delegate_task(goal="Test child status", parent_agent=parent)
+            )
+            return result["results"][0]
+
+    def test_failed_flag_marks_status_failed(self):
+        """Regression (issue #97655): a provider-rejected child (HTTP 400 on its
+        first call) returns completed=False with failed=True + an error string.
+        It must be status=failed, exit_reason=error, and NOT truncated."""
+        entry = self._delegate_single(
+            {
+                "final_response": "HTTP 400: upstage/solar-pro-4 is not a valid model ID",
+                "completed": False,
+                "interrupted": False,
+                "failed": True,
+                "error": "HTTP 400: upstage/solar-pro-4 is not a valid model ID",
+                "api_calls": 1,
+                "messages": [],
+            }
+        )
+        self.assertEqual(entry["status"], "failed")
+        self.assertEqual(entry["exit_reason"], "error")
+        self.assertFalse(entry["truncated"])
+
+    def test_error_with_summary_still_failed(self):
+        """A child that returns BOTH an error field and a summary must still be
+        failed — the summary-presence heuristic must not override the
+        structured failure."""
+        entry = self._delegate_single(
+            {
+                "final_response": "partial work before crashing",
+                "completed": False,
+                "interrupted": False,
+                "failed": True,
+                "error": "provider boom",
+                "api_calls": 3,
+                "messages": [],
+            }
+        )
+        self.assertEqual(entry["status"], "failed")
+        self.assertEqual(entry["exit_reason"], "error")
+        self.assertFalse(entry["truncated"])
+
+    def test_genuine_truncation_stays_completed_max_iterations(self):
+        """REGRESSION GUARD: a child that genuinely exhausts its iteration
+        budget (completed=False, no failed flag, no error) but still returns a
+        summary must keep status=completed, exit_reason=max_iterations, and
+        truncated=True. This is the legitimate truncation path we must not
+        break while making failure labels honest."""
+        entry = self._delegate_single(
+            {
+                "final_response": "made partial progress before the budget ran out",
+                "completed": False,
+                "interrupted": False,
+                "api_calls": 10,
+                "messages": [],
+            }
+        )
+        self.assertEqual(entry["status"], "completed")
+        self.assertEqual(entry["exit_reason"], "max_iterations")
+        self.assertTrue(entry["truncated"])
+
+    def test_interrupted_unchanged(self):
+        """Interrupted children keep status=interrupted + exit_reason=interrupted
+        and are not marked truncated."""
+        entry = self._delegate_single(
+            {
+                "final_response": "some partial output",
+                "completed": False,
+                "interrupted": True,
+                "api_calls": 2,
+                "messages": [],
+            }
+        )
+        self.assertEqual(entry["status"], "interrupted")
+        self.assertEqual(entry["exit_reason"], "interrupted")
+        self.assertFalse(entry["truncated"])
+
+
 class TestSubagentCostRollup(unittest.TestCase):
     """Port of Kilo-Org/kilocode#9448 — parent's session_estimated_cost_usd
     must include subagent spend, not just the parent's own API calls."""
--- a/tests/tools/test_delegate.py
+++ b/tests/tools/test_delegate.py
@@ -741,6 +741,47 @@ class TestDelegateFailedChildStatus(unittest.TestCase):
         self.assertEqual(entry["exit_reason"], "error")
         self.assertFalse(entry["truncated"])
 
+    def test_error_without_failed_flag_marks_failed(self):
+        """A child result that carries a non-empty error string but OMITS the
+        ``failed`` key entirely (not ``failed=False`` — the key is absent, as in
+        legacy/partial result dicts) must still be status=failed + exit_reason=error.
+        The status branch checks ``result.get('failed') or result.get('error')``,
+        so the error field alone has to win — otherwise a dropped ``failed`` key
+        would silently mislabel a provider rejection as budget exhaustion."""
+        entry = self._delegate_single(
+            {
+                "final_response": "connection reset while streaming",
+                "completed": False,
+                "interrupted": False,
+                "error": "connection reset",
+                "api_calls": 2,
+                "messages": [],
+            }
+        )
+        self.assertEqual(entry["status"], "failed")
+        self.assertEqual(entry["exit_reason"], "error")
+        self.assertFalse(entry["truncated"])
+
+    def test_empty_error_with_summary_is_completed(self):
+        """REGRESSION PIN: an empty-string ``error`` field must NOT be treated as
+        a failure. ``result.get('error')`` returns ``''`` which is falsy, so the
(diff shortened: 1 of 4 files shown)


alt-glitch:

Duplicate of #97665 for the child failure-status repair. The config-level model-rejection notice also overlaps #97667.

DavidMetcalfe:
@alt-glitch Fact-checked the duplicate label against both cited PRs — the overlap is real, but there are distinct scope differences outlined below. Leaving the label decision to the maintainers.

Both #97665 (@fangliquanflq) and #97667 (@liuhao1024) are sound work addressing #97655 and #97654. I ran mutation checks on both heads to confirm their regression tests actually bite: removing #97665's `terminal_failure` branch fails its test (`test_provider_failure_is_not_reported_as_iteration_exhaustion`), disabling #97667's notice function fails `test_batch_model_rejection_notice_prepended`, and both restore green. Also worth noting: both PRs were opened within ~25 minutes of the issues being filed, so this is independent parallel work, not copying.

Scope differences, for whoever picks the winners:

**vs #97665** (status fix): theirs additionally skips schema-retry on terminal results. Mine additionally carries the `truncated`-flag regression guard for genuine budget exhaustion, edge-case tests (error present without a `failed` key; empty `error` string with a summary must stay `completed`), and the `exit_reason` enum documentation.

**vs #97667** (rejection notice): theirs covers the batch renderer only; mine covers both the batch and single-dispatch renderers, and adds the "No fallback chain is configured, so no failover was attempted" line with its test (the default install has no chain, so the issue's failover hint is otherwise misleading).

Not asking for label removal — if maintainers prefer merging the earlier PRs, the unique deltas above are portable in either direction. One possible combination: #97667 for the batch renderer plus this PR's single-dispatch coverage, with either implementation of #97655 (theirs has the schema-retry skip, mine has the edge-case tests).

teknium1:
Thanks @DavidMetcalfe — this was the right fix, and more thorough than our first pass. All three of your commits were cherry-picked onto current main with your authorship preserved and merged in #99067 (merge commit cd2bd160579d). Your honest exit_reason/status classification now ships together with the user-facing failure notices from #99049 (CLI tree line + gateway chat notice even with tool_progress off). Closing this PR since the work is on main under your name.

PR #67523, opened 2026-07-19
feat(desktop): add guided subagent model/provider picker (#67347)

webtecnica:
## Summary

Implements #67347: Adds a guided provider + model picker for `delegation.model` and `delegation.provider` in **Desktop** Settings → Advanced, replacing the two bare free-text inputs with a paired dropdown sourced from the gateway's model catalog.

## Changes

| File | Change |
|------|--------|
| `apps/desktop/src/app/settings/delegation-model-provider-field.tsx` | New component — paired provider + model picker mirroring FallbackModelsField pattern |
| `apps/desktop/src/app/settings/delegation-model-provider-field.test.tsx` | Vitest tests covering inherit mode, provider selection, free-text model input for custom endpoints, external config resync |
| `apps/desktop/src/app/settings/config-settings.tsx` | Inline hook: when schema key is `delegation.model`, render the paired picker instead of two bare inputs; `delegation.provider` is excluded from the field list |
| `apps/desktop/src/i18n/en.ts` | English strings: `delegationInheritFromMain`, `delegationInheritModel`, `delegationCustomModelId` |
| `apps/desktop/src/i18n/types.ts` | Type definitions for new i18n keys |

## Design

- **Provider dropdown** sourced from `getGlobalModelOptions()` — same data source as the chat model picker and fallback-models field
- **Model dropdown** filtered by selected provider, with free-text fallback for custom model IDs not in any catalog
- **"Inherit from main agent"** surfaced as a first-class dropdown option that writes `""` to both keys
- **Atomic write** of both config keys on every change — transient invalid pairs are never persisted
- **Custom-endpoint handling**: providers with empty `models` lists show a free-text input automatically
- Mirrors the established `FallbackModelsField` pattern (owned local state, external-resync via `useRef`)

## Verification

- [x] Component renders "Inherit from main agent" when both fields are empty
- [x] Selecting a provider shows its models in a dropdown
- [x] Custom endpoint providers show free-text model input
- [x] Switching to "Inherit from main agent" clears both fields atomically
- [x] External config changes (profile switch) correctly resync local state

Closes #67347

PR #98130, opened 2026-08-29
feat(config): write-time model-slug validation in hermes config set

### LINKED ISSUE #97656: [Bug]: hermes config set accepts nonexistent model slugs with no catalog validation

DavidMetcalfe:
## What does this PR do?

Adds write-time validation to `hermes config set` for model-routing keys
(`model.default`, bare `model` shorthand, `delegation.model`,
`auxiliary.<task>.model`): the slug is checked against the *same cached
provider catalog the model picker uses* and a warning + confirmation fires
when it isn't present.

## Motivation

Today any string is accepted silently. A real incident: a one-hyphen typo
(`upstage/solar-pro-4` vs the valid `upstage/solar-pro4`) was written with a
success message and caused a week of instantly-failing delegations whose cause
was buried in batch reports. The UI half of this gap is covered by guided
pickers (#67523/#67557/#76480); this covers CLI, scripted, and agent-written
config, which the pickers don't reach.

## Behavior

| Case | Result |
|---|---|
| Known slug | no warning, proceeds |
| Unknown slug, interactive TTY | warning (stderr) + closest matches (difflib, ≤3) + `Set anyway? [y/N]`; N aborts (exit 1) |
| Unknown slug, non-TTY | warning to stderr, proceeds (scripts/agents keep working) |
| `--force` | warning printed, prompt skipped |
| Provider empty/`auto`/`custom`/unknown, or any validation exception | fail open — write proceeds untouched |

Details:

- Provider pairing per key: `model.default`→`model.provider`,
  `delegation.model`→`delegation.provider`,
  `auxiliary.<task>.model`→`auxiliary.<task>.provider` — with fallback to the
  parent `model.provider` when the leaf's own provider is empty (inherit).
- Catalog reuse: `cached_provider_model_ids()` — the picker's existing
  disk-cached catalog; no new network layer, and a cold cache triggers the
  picker's existing probe.
- OpenRouter `:nitro`-style routing suffixes match on their base id (gated to
  openrouter only).
- Interactive detection requires both stdin and stdout TTYs.
- Setting a *provider* key re-checks its already-configured sibling model, so
  the provider-then-model ordering gap (model written first, unvalidated
  because provider was unknown) closes on the very next provider write.

## Changes Made

- `hermes_cli/config.py` — `_model_routing_provider_key` (with the
  `model.default` leaf-guard fix), `_provider_for_config_path` (+ parent
  fallback), `_slug_in_catalog`, `_validate_model_slug_for_write`, provider-set
  sibling re-check; gate called before `_set_nested` so an abort writes
  nothing.
- `tests/hermes_cli/test_config_model_slug_validation.py` — 19 tests, network
  isolated (catalog mocked).

## How to Test

1. `hermes config set delegation.model totally-fake-model-xyz` in a terminal →
   warning + prompt; N aborts.
2. Same piped/non-TTY → warning only, write proceeds.
3. `hermes config set model.default gpt-5.5 --provider-openai` equivalent
   writes → no warning for real catalog slugs.
4. Any network/cache failure → write proceeds as before (fail open).

## Review

Two rounds of cross-vendor review (Gemini 3.7 Flash High + GPT-OSS 120B
Medium, full diffs, byte-identical briefs). Round-2 verdicts: ACCEPTABLE from
both reviewers ("ready to merge"). Notable round-1 dispositions verified
against source: the `model.default`-never-validated bug (Flash BLOCKER — real,
refined: the original parse-failure claim was wrong but the leaf guard was),
provider-inheritance fallback (real, fixed), GPT-OSS's "config load masking"
BLOCKER rejected as contradicting the issue's own fail-open contract.
Commit bodies carry the full per-finding ledger. 139 passed post-rebase on
target suites.

Closes #97656

### DIFF
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -5588,7 +5588,11 @@ def _model_routing_provider_key(key: str) -> Optional[str]:
     catalog the model id must come from.
     """
     segs = key.strip().lower().split(".")
-    if not segs or segs[-1] != "model":
+    if not segs:
+        return None
+    # Model-routing leaf keys are ``<...>.model`` plus the ``model`` /
+    # ``model.default`` dual (the mapping stores the id at ``model.default``).
+    if segs[-1] != "model" and segs != ["model", "default"]:
         return None
     if segs in (["model"], ["model", "default"]):
         return "model.provider"
@@ -5599,20 +5603,37 @@ def _model_routing_provider_key(key: str) -> Optional[str]:
     return None
 
 
+def _config_node(container: Any, path: str) -> Any:
+    """Walk *container* by dotted *path*, returning the node or ``None``."""
+    node = container
+    for seg in path.split("."):
+        if not isinstance(node, dict):
+            return None
+        node = node.get(seg)
+    return node
+
+
 def _provider_for_config_path(provider_key: str) -> str:
-    """Read a provider config path from the merged config as a stripped string."""
+    """Read a provider config path from the merged config as a stripped string.
+
+    ``delegation.provider`` and ``auxiliary.<task>.provider`` may be *explicitly*
+    set to ``''`` to mean "inherit the parent provider" (delegation defaults to
+    ``''``; auxiliary tasks default to the ``auto`` sentinel).  An empty leaf
+    therefore falls back to the global ``model.provider`` so a delegation /
+    auxiliary model slug is validated against the provider that actually
+    resolves it.  ``model.provider`` is the root of the chain and has nothing
+    to fall back to.
+    """
     try:
         cfg = load_config()
     except Exception:
         return ""
-    node: Any = cfg
-    for seg in provider_key.split("."):
-        if not isinstance(node, dict):
-            return ""
-        node = node.get(seg)
-    if isinstance(node, str):
-        return node.strip()
-    return ""
+    node = _config_node(cfg, provider_key)
+    value = node.strip() if isinstance(node, str) else ""
+    if not value and provider_key != "model.provider":
+        parent = _config_node(cfg, "model.provider")
+        value = parent.strip() if isinstance(parent, str) else ""
+    return value
 
 
 def _provider_skips_catalog_validation(provider: str) -> bool:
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #97660: both implement catalog-aware warning and confirmation for invalid model-routing slugs in `hermes config set`.

PR #95704, opened 2026-08-26
fix(desktop): don't kill a live macOS backend on parent-marker timezone drift

ygd58:
Fixes #95693.

## Root cause

`_process_start_marker()`'s macOS fallback fingerprints the parent Electron process with `ps -p <pid> -o lstart=` -- a timezone/locale-rendered wall-clock string, not a stable identity value like the `linux:`/`win:` markers other platforms produce. Electron caches its marker once for the app's lifetime; the Python backend re-probes on every spawn. If the system timezone changes while Electron stays alive, the two renderings of the SAME instant diverge byte-for-byte, `_parent_start_markers_match()` reports a mismatch, and `_is_serve_orphaned()` treats it as definitive proof of death -- `os._exit(0)` fires within ~5ms of boot, silently (bypasses atexit/logging flush).

`_is_serve_orphaned()`'s own docstring already promises fail-safe behavior for inconclusive probes. A `ps:` marker mismatch is exactly that case -- it's a rendering artifact, not identity -- but was being treated as positive proof instead.

## Fix

Implemented the issue's own suggested fix #3 (safest, lowest-risk -- no cross-language coordination with the Electron/TypeScript side needed): when both markers use the `ps:` prefix and don't match, degrade to the PID-only liveness check instead of declaring the backend orphaned. A mismatch between stable formats (`linux:`/`win:`) still correctly proves a dead/recycled parent and is untouched -- scoped specifically to `ps:`.

Also implemented suggested fix #4: a warning log line before `os._exit(0)`, naming the desktop PID and expected marker -- the issue calls the current silent exit "the single biggest cost of this bug."

## Verification

Verified directly against the issue's own live-instance evidence (the exact EDT vs CEST marker pair for the same instant) -- confirms the fix correctly returns `False` for a genuinely-alive parent. Also verified: a genuinely dead parent behind a `ps:` mismatch is still detected; an exact `ps:` match still short-circuits; a `linux:` marker mismatch (recycled PID) still correctly returns `True`.

Added 4 regression tests to the existing test file. Verified as a genuine regression by reverting just the fix and confirming the exact-evidence test fails.

9/9 pass in the modified test file; 3/3 in a related boot-handshake test file (no regression).

PR #93990, opened 2026-08-24
fix(desktop): tolerate macOS parent marker drift

Ahmett101:
Summary
Desktop local mode uses a parent-death watchdog so a backend exits when the Electron owner is gone. On macOS both Electron and Python use the POSIX `ps -o lstart=` fallback marker, which is a presentation string; spacing/locale drift can make the backend treat a live Desktop parent as PID reuse and cleanly exit with code 0 immediately after announcing its port. This change keeps exact matches as before but treats `ps:` marker mismatches as inconclusive, degrading to fail-safe liveness behavior instead of killing a healthy backend, while preserving conclusive Linux/Windows machine-marker mismatches.

Changes
hermes_cli/web_server.py: route parent-start marker mismatches through a conclusive/inconclusive classifier; `ps:` mismatches now fail safe instead of exiting the backend.
tests/hermes_cli/test_serve_parent_watchdog.py: adds regression coverage for macOS `ps:` marker drift and confirms Linux-style machine-marker mismatches still protect against PID reuse.

How to Test
/data/data/com.termux/files/home/.hermes/hermes-agent/venv/bin/python -m pytest tests/hermes_cli/test_serve_parent_watchdog.py -q
# ✅ 7/7 passed
ruff check hermes_cli/web_server.py tests/hermes_cli/test_serve_parent_watchdog.py
# ✅ passed
git diff --check
# ✅ passed
python scripts/check-windows-footguns.py --diff upstream/main
# ✅ passed

Checklist
- [x] Tests pass — 7/7 targeted
- [x] Follows Conventional Commits
- [x] Changes scoped to this fix only
- [x] Cross-platform impact assessed (Linux / macOS / WSL2 / Windows / Termux)
- [x] profile-safe paths used — no hardcoded ~/.hermes
- [x] .env not used for non-credential settings (behavioral settings → config.yaml)

Risk & Impact
Low. The change only relaxes non-machine `ps:` marker mismatches to fail safe; Linux and Windows machine markers still conclusively detect PID reuse.
Type: Bug fix
Closes: #93958

PR #98132, opened 2026-08-29
fix(desktop): ignore truncated macOS ps parent markers

drewTuzson:
## Summary
On macOS, Desktop-spawned `hermes serve` can print `HERMES_BACKEND_READY` and then `os._exit(0)` because the parent-death watchdog treats a truncated or padded `ps:` start marker as a dead owner.

Reproduced locally: `HERMES_PARENT_START_MARKER=ps:Sat` (or a padded `ps:Sat Aug 29 15:04:31 2026    ` mismatch) exits 0 immediately after ready. A full matching `ps:` lstart stays up.

This made Hermes Desktop show `Hermes couldn't start` / `The background gateway didn't come up` even though Electron was still alive.

## Change
- Require a full `ps -o lstart=` value (year + time), so `ps:Sat` is invalid and the watchdog no-ops instead of killing the backend.
- Normalize `ps:` whitespace before compare.
- Treat empty inherited marker/nonce env as absent.
- Log before the watchdog `os._exit(0)`.

## Test
`python -m pytest tests/hermes_cli/test_serve_parent_watchdog.py -o addopts= -q` (6 passed)

No Electron rebuild is required. Packaged Desktop already spawns the git-install Python backend.

### DIFF
--- a/hermes_cli/web_server.py
+++ b/hermes_cli/web_server.py
@@ -221,13 +221,30 @@ def _process_start_marker(pid: int) -> str:
     raise OSError(f"ps could not inspect PID {pid}: {result.stderr.strip()}")
 
 
+def _normalize_parent_start_marker(marker: str) -> str:
+    """Collapse ps(1) column padding so Node trim and Python strip compare equal."""
+    prefix, separator, value = marker.partition(":")
+    if not separator:
+        return marker
+    return f"{prefix}:{' '.join(value.split())}"
+
+
 def _valid_parent_start_marker(marker: str) -> bool:
     prefix, separator, value = marker.partition(":")
     if not separator or not value or value != value.strip():
         return False
     if prefix in ("linux", "win", "winms"):
         return value.isdigit()
-    return prefix == "ps"
+    if prefix != "ps":
+        return False
+    # macOS/BSD ``ps -o lstart=`` looks like ``Sat Aug 29 15:04:31 2026``.
+    # A truncated token (``ps:Sat``) is what you get if the marker is split on
+    # spaces in the environment; treating that as a real identity makes a live
+    # Desktop look dead and the backend exits 0 right after HERMES_BACKEND_READY.
+    tokens = value.split()
+    has_year = any(token.isdigit() and len(token) == 4 for token in tokens)
+    has_time = any(":" in token for token in tokens)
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Fixes Desktop-spawned `hermes serve` exiting immediately after `HERMES_BACKEND_READY` on macOS when parent marker is truncated/padded, causing "Hermes couldn't start" false alarm. Small, targeted.

**What it does**
- `hermes_cli/web_server.py:_normalize_parent_start_marker` collapses `ps(1)` column whitespace (`' '.join(value.split())`) so Node `trim` vs Python `strip` compare equal for `ps:` markers.
- `_valid_parent_start_marker` for `ps:` now requires `len(tokens)>=4` and presence of a 4-digit year and `:` time token — rejects truncated `ps:Sat` (split on spaces in env, or truncated env var) that previously passed as "valid" and caused `_is_serve_orphaned` to treat a live Desktop as dead.
- `_parent_start_markers_match` tries normalized equality before `win:`/`winms:` legacy check.
- `_start_parent_death_watchdog` normalizes env reads (`or None`), logs `parent-death watchdog: owner is gone; exiting 0 (pid=... marker=...)` before `os._exit(0)` for observability.
- Test `test_macos_ps_marker_requires_full_lstart_not_a_truncated_weekday` pins: full `ps:Sat Aug 29 15:04:31 2026` valid, `ps:Sat` invalid, padded match true, `_is_serve_orphaned` with padded actual false.

**Non-blocking notes**
- Validation `len>=4 && has_year && has_time` correctly rejects `ps:Sat` and accepts full `lstart`. Edge: localized `ps` output with non-English weekday/month still has year/time, so passes; alternative BSD locales still produce year/time, so robust.
- `or None` for env vars treats empty string as missing — correct; previously `""` would be processed as marker, now treated as `None` and watchdog skips.
- Print before `_exit` uses `flush=True` — ensures log reaches Desktop before process dies even if stdout buffered.

Minimal risk.

austinpickett:
Superseded by #103172, which consolidates this with the other two open fixes for the same `ps:` parent-marker path (#95704, #93990, #98132) and carries your authorship via Co-authored-by. Thanks for the fix and the reproduction.

Issue #29344, opened 2026-05-20
[Bug]: _is_entitlement_failure over-matches xAI 'bad-credentials' 403 — long-running TUI sessions can't auto-refresh stale OAuth tokens


SmelterLabs:
## Summary

`_is_entitlement_failure` in `run_agent.py` over-matches on xAI Grok 403 responses, causing legitimate "OAuth access token failed validation" errors to be misclassified as unsubscribed-account entitlement failures. The defensive guard against entitlement refresh loops (existing test references issue #26847) suppresses the refresh-on-401 path for both real cases, leaving long-running TUI sessions stuck on a stale token with no recovery.

Workaround: exit and reopen the TUI — the startup refresh path bypasses the broken classifier.

## Repro

1. Open a Hermes TUI session against `provider/xai-oauth` (SuperGrok).
2. Let it sit idle long enough that the access token goes stale by xAI's server-side criteria (in my case, ~22 hours; can happen sooner if xAI rotates session-side).
3. Send a request.
4. xAI returns HTTP 403 with this body:

```json
{
  "code": "The caller does not have permission to execute the specified operation",
  "error": "The OAuth2 access token could not be validated. [WKE=unauthenticated:bad-credentials]"
}
```

5. Hermes logs `Non-retryable client error` and surfaces it to the user. No refresh attempt happens, even though the credential pool's `_refresh_entry` for this provider works fine (proven by opening a new TUI session — the startup-resolve path refreshes successfully).

## Expected

The `[WKE=unauthenticated:bad-credentials]` suffix unambiguously indicates this is a credential-validation failure, not an entitlement failure. Hermes should:
- Call `_recover_with_credential_pool` → `try_refresh_current()` → `_swap_credential`
- Retry the request with the refreshed token
- Either succeed (the typical case after a stale token) or, if the refresh itself fails terminally, fall through to the existing terminal-quarantine path

## Actual

`_is_entitlement_failure` returns `True` because the response body matches its substring heuristic on `"caller does not have permission"`. The recovery short-circuits, returns `False`, error surfaces as non-retryable.

## Root cause

xAI's API returns the *same* `code` field text for two distinct conditions:

| Condition | `code` (same) | `error` field (the disambiguator) |
|---|---|---|
| Entitlement (account isn't SuperGrok-subscribed) | `"The caller does not have permission to execute the specified operation"` | `"... active Grok subscription. Manage at https://grok.com"` (or similar entitlement language) |
[...]

PR #98133, opened 2026-08-29
fix(agent): refresh xai-oauth on its 403 token-expiry signal, not 401 only

slawt:
## Problem

xAI expires an OAuth access token with **HTTP 403** — carrying either `[WKE=unauthenticated:…]` or `OAuth2 access token could not be validated` — and never sends a 401.

The singleton `codex_responses` refresh branch in `run_conversation` gates on `status_code == 401`:

```python
if (
    agent.api_mode == "codex_responses"
    and agent.provider in {"openai-codex", "xai-oauth"}
    and status_code == 401          # xAI never sends this
    and not _retry.codex_auth_retry_attempted
):
```

so that arm cannot fire for `xai-oauth` at all.

The credential-pool arm already draws this distinction (#29344 deliberately exempts those two bodies from the entitlement override), but `recover_with_credential_pool()` returns early whenever `agent._credential_pool is None`.

With both arms out, a long-running gateway on `xai-oauth` keeps re-presenting a token that died ~6h earlier with no path back. Grok stops answering until the process is restarted — and since `hermes update` restarts the stack and re-mints, updating appears to "fix Grok" for exactly one token lifetime at a time.

## Evidence

From a gateway host running `grok-4.6` on `xai-oauth`:

```
agent.conversation_loop: API call failed (attempt 1/3)
  error_type=PermissionDeniedError provider=xai-oauth
  base_url=https://api.x.ai/v1 model=grok-4.6
  summary=HTTP 403: {"code":"unauthenticated:bad-credentials",
                     "error":"The OAuth2 access token could not be validated."}
```

Recurring across four consecutive days. At the time of capture the access token (6h lifetime) sat **16h past `exp`** with `last_refresh` 22h stale, and there were **zero** credential-pool log lines in that window — consistent with the silent `pool is None` early return.

The refresh token was healthy throughout: a single forced `resolve_xai_oauth_runtime_credentials(force_refresh=True)` minted a fresh 6h token immediately and chat recovered. The credential was never dead — nothing was pulling the trigger.

## Fix

Lift the #29344 predicate into `is_xai_stale_oauth_error()` and call it from both arms so they cannot drift, then widen the singleton gate to `401 or <stale xai 403>`.

Entitlement 403s still classify as entitlement, so the refresh loop cannot spin on an unsubscribed account — that guard is what the shared predicate exists to protect.

## Tests

`tests/agent/test_xai_stale_oauth_403_refresh.py` — 15 cases: the production body verbatim, the `WKE` variant, casing, entitlement and org-policy 403s that must **not** refresh, non-403 statuses, and degenerate input. Plus two structural tests pinning the widened gate and the single-owner predicate.

All three structural tests fail against unpatched `main`. Existing suites stay green:

```
tests/run_agent/test_codex_xai_oauth_recovery.py
tests/agent/test_credential_pool_routing.py
tests/agent/test_custom_pool_mismatch_guard.py
tests/agent/test_billing_unverified_carrythrough.py
tests/hermes_cli/test_auth_xai_oauth_provider.py
                                        72 passed
```

## Note

This only restores the *reactive* path. The pool arm being skipped when `_credential_pool is None` is a separate question — worth a look, but out of scope here.

### DIFF
--- a/agent/agent_runtime_helpers.py
+++ b/agent/agent_runtime_helpers.py
@@ -1114,14 +1114,19 @@ def is_xai_stale_oauth_error(
     )
 
 
+_DEVICE_CODE_SOURCES = frozenset({"device_code", "manual:device_code"})
+
+
+def oauth_pool_entry_is_device_code(entry) -> bool:
+    return str(getattr(entry, "source", "") or "") in _DEVICE_CODE_SOURCES
+
+
 def oauth_active_key_is_foreign_pool_entry(agent, active_key: str) -> bool:
     """True when ``active_key`` belongs to a pooled credential.
 
     Used by the singleton OAuth refresh guard: if the live key is a real
     pool entry, force-refreshing/adopting the device_code singleton would
-    silently swap accounts. If the live key matches *no* entry, it is a
-    stale in-memory copy (the pool/store already rotated) and adopting
-    the singleton is the recovery, not an account swap.
+    silently swap accounts.
     """
     if not active_key:
         return False
@@ -1138,6 +1143,34 @@ def oauth_active_key_is_foreign_pool_entry(agent, active_key: str) -> bool:
         if runtime and runtime == needle:
             return True
     return False
+
+
+def resolve_stale_oauth_pool_entry(agent, pool):
+    """Pool entry that owns a stale-token failure, or None if we must not guess.
+
+    ``_credential_pool_entry_id`` survives access-token rotation, so it is
+    preferred even when ``runtime_api_key`` no longer matches. Without an
+    id, only a sole device_code entry is unambiguous — a sole MANUAL
+    other-account entry is not the failed request's identity.
+    """
+    if pool is None:
+        return None
+    try:
+        entries_fn = getattr(pool, "entries", None)
+        entries = list(entries_fn() or []) if callable(entries_fn) else []
+    except Exception:
+        return None
+    by_id = {}
+    for entry in entries:
+        eid = getattr(entry, "id", None)
+        if isinstance(eid, str) and eid:
+            by_id[eid] = entry
+    cred_id = getattr(agent, "_credential_pool_entry_id", None)
+    if isinstance(cred_id, str) and cred_id and cred_id in by_id:
+        return by_id[cred_id]
+    if len(entries) == 1 and oauth_pool_entry_is_device_code(entries[0]):
+        return entries[0]
+    return None
 
 
 def recover_with_credential_pool(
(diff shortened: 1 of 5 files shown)


Enough1122:

**Overall:** Fixes `xai-oauth` refresh never firing because xAI uses 403 (not 401) for expired tokens. Centralizes predicate so credential-pool and singleton `codex_responses` arms cannot drift. Explains why `hermes update` appeared to "fix Grok" for one lifetime.

**What it does**
- `agent/agent_runtime_helpers.py` adds `is_xai_stale_oauth_error(error_context, status_code)` (403-only, scans `message|reason|code|error` lowercased for `[wke=unauthenticated:` or `oauth2 access token could not be validated`) — shared by both paths, case-insensitive.
- Refactors `recover_with_credential_pool` to call the shared helper instead of inline `_is_xai_auth_failure`.
- `agent/conversation_loop.py` widens singleton refresh gate from `status_code==401` to `status_code==401 or _xai_stale_oauth` for `agent.provider=="xai-oauth"`; `_xai_stale_oauth` computed via helper. On success, vprint now shows actual status (`after {_seen}`) instead of hardcoded 401.
- Tests `test_xai_stale_oauth_403_refresh.py`: parametric predicate (real expired body, WKE, OAuth2 phrase, uppercase, entitlement, org not allowed, wrong status, degenerate inputs), source-text guard that gate contains `_xai_stale_oauth`, and pool-arm source check that it calls helper and no inline duplicate.

**Non-blocking notes**
- Predicate scans `error_context` dict values, not raw HTTP body string — `error_context` is built upstream from the error; ensure the 403 body is always mapped into one of `message|reason|code|error` keys. Real log body `{"code":"unauthenticated:bad-credentials","error":"The OAuth2..."}` would map to `error` and `code` — covered.
- Gate keeps `agent.provider in {"openai-codex","xai-oauth"}` — `openai-codex` still 401-only (correct; only `xai-oauth` 403 is stale-token). `_xai_stale_oauth` explicitly requires `agent.provider=="xai-oauth"`, so no cross-provider false positive.
- `_is_xai_auth_failure` removal verified by test (`assert "_is_xai_auth_failure" not in recover`) — prevents drift.

Correct fix.

slawt:
Follow-up on this branch: the 403 gate was necessary but not sufficient.

The 8:34 PM failure still died after this PR because the expired in-memory access token matched **no** pool entry (`credential pool: failed credential identity matched no xai-oauth entry`). `try_refresh_matching(hint=expired)` returned None, the single-entry pool refused to rotate, and the singleton guard skipped because store token ≠ live token (it treated that as an account swap).

This push:

- **Pool arm:** on an xAI stale 403, if the hint misses and the pool has one entry, refresh that entry.
- **Singleton arm:** if the live key is not a pooled credential, adopt the current store token instead of skipping. Explicit `api_key=` / other-account (no pool, or live key *is* a pool entry) still skips.

Tests added in `test_xai_stale_oauth_403_refresh.py` and `test_run_agent_codex_responses.py`.

slawt:
Independent review of the identity-miss follow-up: the first patch could swap onto the device_code singleton after a MANUAL remint, and the pool arm force-refreshed (spending the new single-use refresh token) instead of adopting.

This push gates recovery on `_credential_pool_entry_id` or a sole device_code entry, adopts a reminted token without a second POST, and only adopts the store singleton when that JWT is not expired.

PR #95235, opened 2026-08-26
fix(desktop): avoid unnecessary Linux SUID sandbox setup

EmpireOperating:
## Summary
- Skip privileged `chrome-sandbox` ownership repair on Linux hosts where Chromium can use its normal user-namespace sandbox.
- Detect kernel policies that disable user namespaces, and on Ubuntu AppArmor-restricted hosts run a bounded throwaway namespace probe so profiles explicitly granted `userns` are not treated as blocked.
- Preserve the existing SUID-helper and last-resort `--no-sandbox` behavior only for hosts that actually cannot create a user namespace.

## Why
The unpacked desktop bundle is rebuilt beneath the user home directory during self-update, so `chrome-sandbox` naturally loses `root:root 4755`. Requiring that privileged repair on every Linux launch makes normal self-updates sudo-gated even where Electron's secure namespace sandbox is available.

## Validation
- `env -u ELECTRON_OZONE_PLATFORM_HINT /home/quatro/.hermes/hermes-agent/venv/bin/python -m pytest tests/hermes_cli/test_gui_command.py tests/hermes_cli/test_linux_desktop_entry.py -q` → 59 passed, 7 skipped
- `ruff check hermes_cli/main.py tests/hermes_cli/test_gui_command.py`
- `python -m compileall -q hermes_cli/main.py tests/hermes_cli/test_gui_command.py`
- `hermes desktop --build-only --force-build` produced the Linux unpacked Electron artifact successfully.

## Evidence
Confirmed current upstream `main` unconditionally invokes `_desktop_linux_sandbox_fixup()` before every packaged Linux desktop launch. On the validating Arch host: `unprivileged_userns_clone=1`, `max_user_namespaces=62267`, AppArmor restriction absent, and `unshare --user --map-root-user true` succeeds.

PR #89349, opened 2026-08-18
fix(cli): use Chromium's namespace sandbox when userns is available instead of requiring setuid chrome-sandbox + sudo

4dlt:
## What & why

On Linux, `hermes desktop` refuses to launch unless `chrome-sandbox` in the unpacked Electron app is root-owned with mode 4755, and shells out to `sudo chown/chmod` to make it so. Two failure modes follow:

1. **Silent death from the `.desktop` entry** (#88032, #51327): a desktop-icon launch has no TTY, so `sudo` cannot prompt, the fixup fails, and the launcher exits without ever showing a window or an error.
2. **Every update re-breaks the app** (#58593): each rebuild recreates `chrome-sandbox` owned by the user, and the update hand-off's `linux_gate` in `scripts/desktop-update/posix.sh` then refuses to auto-relaunch ("its sandbox helper needs root ownership"), so the app stays dead until the user runs something from a terminal.

The setuid helper is only Chromium's **fallback** sandbox: on hosts where unprivileged user namespaces work (default Arch, Fedora, most non-hardened distros), Chromium uses the namespace sandbox and never consults `chrome-sandbox`. Requiring root ownership there demands a privilege Chromium won't use — and `posix.sh` already acknowledges this for the *absent*-helper case ("chrome-sandbox ABSENT is fine (namespace-sandbox build; nothing to block on)"); this PR makes the *present-but-user-owned* case consistent with that.

## How

- New `_desktop_linux_userns_sandbox_available()` in `hermes_cli/main.py`: probes the real capability with `unshare --user --map-root-user true` (5s timeout, fails closed on any error, missing binary, or non-Linux).
- `_desktop_linux_sandbox_fixup()` returns success **before the sudo path** when the probe passes. An already root-owned 4755 helper still short-circuits first, and restricted hosts (userns disabled, or Ubuntu 23.10+ AppArmor `apparmor_restrict_unprivileged_userns=1` — where the probe fails with EPERM) keep the existing sudo → `--no-sandbox`-fallback behavior unchanged.
- `linux_gate()` in `scripts/desktop-update/posix.sh` gets the same probe, so the post-update auto-relaunch fires on userns-capable hosts.

Sandboxing stays fully enabled in both regimes; this never adds `--no-sandbox`.

## How to test

[...]

PR #98135, opened 2026-08-29
fix(desktop): no-sudo Linux namespace sandbox for chrome-sandbox 0755

### LINKED ISSUE #51327: [Bug]: Hermes Desktop silently fails from .desktop launcher when Electron chrome-sandbox lacks setuid (4755)
## Bug Description

On Linux, clicking the Hermes Desktop icon (`.desktop` launcher) can fail **silently** — no window, no error dialog. The app aborts immediately because Electron's `chrome-sandbox` helper is not configured as `root:root` with mode `4755`.

This happens when launching via `~/.local/bin/hermes-desktop-launch` (used by the `.desktop` entry). That launcher `exec`s the packaged binary directly and **never runs** the sandbox fixup that `hermes desktop` performs via `_desktop_linux_sandbox_fixup()`.

After a desktop rebuild/update (e.g. `hermes update`, `hermes desktop --force-build`, or interrupted `npm run pack`), `chrome-sandbox` is recreated with normal user permissions (`-rwxr-xr-x pamnard:pamnard`), and the icon launcher breaks until the user manually fixes permissions or discovers `--no-sandbox`.

## Steps to Reproduce

1. Install Hermes Desktop on Linux and create the `.desktop` launcher (`~/.local/share/applications/hermes-desktop.desktop` → `Exec=~/.local/bin/hermes-desktop-launch`).
2. Rebuild the packaged desktop app so `apps/desktop/release/linux-unpacked/` is recreated (update, force-build, or interrupted pack).
3. Confirm sandbox helper has wrong permissions:
   ```bash
   ls -la ~/.hermes/hermes-agent/apps/desktop/release/linux-unpacked/chrome-sandbox
   # -rwxr-xr-x user:user  (bad)
   ```
4. Click the Hermes icon in the app grid — nothing appears.
5. Run the binary directly to see the fatal error:
   ```bash
   ~/.hermes/hermes-agent/apps/desktop/release/linux-unpacked/Hermes 2>&1 | head -5
   ```
   Output:
   ```
   FATAL:sandbox/linux/suid/client/setuid_sandbox_host.cc:166] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that .../chrome-sandbox is owned by root and has mode 4755.
   ```
6. Compare with CLI launch path — `hermes desktop` runs `_desktop_linux_sandbox_fixup()` and prompts for sudo before launching (works).

## Expected Behavior

- Launching Hermes Desktop from the `.desktop` icon should work reliably after rebuild/update.
- Sandbox fixup should run on **every** launch path (not only `hermes desktop` CLI), or the launcher should surface a clear error instead of silently dying.
- Ideally fixup is also re-run automatically after desktop rebuild/update (install.sh already does this during install, but not when user clicks icon later).

## Actual Behavior

- `.desktop` icon click → instant silent crash (exit 133 / SIGTRAP), no UI feedback.
- `hermes desktop` from terminal → works (runs sudo fixup first).
- Workaround: `sudo chown root:root .../chrome-sandbox && sudo chmod 4755 .../chrome-sandbox`, or add `--no-sandbox` to `hermes-desktop-launch`.

## Environment

- **OS:** Linux (X11, Ubuntu-based, kernel 6.17)
- **Python:** 3.11.15 (Hermes venv)
- **Hermes:** v0.16.0 (2026.6.5), commit dc467488a75d
- **Component:** Hermes Desktop (Electron), Linux launcher
- **Launcher:** `~/.local/bin/hermes-desktop-launch` → `~/.hermes/hermes-agent/apps/desktop/release/linux-unpacked/Hermes`
- **`.desktop` entry:** `~/.local/share/applications/hermes-desktop.desktop`

## Debug Report

Not uploaded via `hermes debug share` — repro is deterministic from sandbox permissions + launcher code path. Happy to upload on request.

## Root Cause Analysis

- `_desktop_linux_sandbox_fixup()` exists in `hermes_cli/main.py` (~5164) and is called only from `cmd_gui()` before launch (~5382).
- `install.sh` also configures sandbox during desktop build (~2500–2521).
- **`hermes-desktop-launch`** bypasses both — it directly `exec`s the Electron binary with no fixup:
  ```bash
  exec "$HERMES_HOME/hermes-agent/apps/desktop/release/linux-unpacked/Hermes" "$@"
  ```
- After any rebuild that recreates `linux-unpacked/`, permissions reset → icon launcher broken until manual sudo fix.

## Proposed Fix

1. Move sandbox fixup into `hermes-desktop-launch` (call a small helper or inline check + sudo prompt), **or**
2. Have `hermes desktop` / install flow write the launcher to always run through a wrapper that performs fixup, **or**
3. Re-run `_desktop_linux_sandbox_fixup()` after every successful desktop build/update and document that icon launch depends on it.

Minimal change: add fixup logic to `hermes-desktop-launch` mirroring `main.py`, or invoke `hermes desktop` from the `.desktop` file instead of the raw binary.

## Workaround (users)

```bash
sudo chown root:root ~/.hermes/hermes-agent/apps/desktop/release/linux-unpacked/chrome-sandbox
sudo chmod 4755 ~/.hermes/hermes-agent/apps/desktop/release/linux-unpacked/chrome-sandbox
```

Verified: app starts normally with `--no-sandbox`; starts normally after correct 4755 permissions.

kvnloo:
## What & why

Linux `hermes desktop` still requires a root-owned 4755 `chrome-sandbox` and shells out to `sudo`. From a `Terminal=false` `.desktop` / no-TTY CLI launch, sudo cannot prompt, so the launcher `sys.exit(1)`s even when unprivileged user namespaces work (canonical **#51327**).

Existing open PRs (#89349, #90108, #95235, #96687, #97404, #51334, …) cover pieces of this. None on current `main` both (1) skip sudo after a fail-closed `unshare --user --map-root-user` probe **and** (2) pass `--disable-setuid-sandbox` so a *present-but-non-setuid* helper does not abort Chromium. This PR does both. It never adds `--no-sandbox` on the userns path.

## How

- Recovered/rebased 4dlt’s userns probe (`_desktop_linux_userns_sandbox_available`) so `_desktop_linux_sandbox_fixup` returns success before sudo when the probe passes; `linux_gate()` in `scripts/desktop-update/posix.sh` matches.
- After a successful non-4755 fixup, append `--disable-setuid-sandbox` (keeps the namespace sandbox).
- Restricted hosts (probe fails) keep the existing sudo → `--no-sandbox` fallback.

## Test

- `pytest tests/hermes_cli/test_linux_sandbox_fixup.py tests/hermes_cli/test_gui_command.py -q` → 52 passed, 9 skipped
- Live host (CachyOS, `unprivileged_userns_clone=1`, helper `kvn:kvn 0755`): userns probe True; fixup True with no sudo; disable-setuid True.

Fixes #51327

Ready for independent maintainer review following author approval. No merge authority.

### DIFF
--- a/contributors/emails/a.dltorreruiz@gmail.com
+++ b/contributors/emails/a.dltorreruiz@gmail.com
@@ -1,0 +1,2 @@
+4dlt
+# PR #98135 attribution repair
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -8184,6 +8184,29 @@ def _desktop_linux_sandbox_fixup(packaged_executable: Path) -> bool:
     return True
 
 
+def _desktop_linux_needs_disable_setuid_sandbox(packaged_executable: Path) -> bool:
+    """Return True when Chromium should skip the present-but-non-setuid helper.
+
+    A user-owned ``chrome-sandbox`` still makes Chromium abort with
+    ``setuid_sandbox_host`` even when the namespace sandbox works. Passing
+    ``--disable-setuid-sandbox`` keeps the userns sandbox and avoids sudo.
+    Call only after ``_desktop_linux_sandbox_fixup`` succeeded without making
+    the helper root-owned 4755 (the userns path). Does not re-probe userns.
+    """
+    if sys.platform != "linux":
+        return False
+    sandbox = packaged_executable.parent / "chrome-sandbox"
+    try:
+        sandbox_lstat = sandbox.lstat()
+    except OSError:
+        return False
+    if not stat.S_ISREG(sandbox_lstat.st_mode):
+        return False
+    if sandbox_lstat.st_uid == 0 and stat.S_IMODE(sandbox_lstat.st_mode) == 0o4755:
+        return False
+    return True
+
+
 _LINUX_PASSWORD_STORES = frozenset({"gnome-libsecret", "kwallet", "kwallet5", "kwallet6", "basic"})
 
 
@@ -8578,6 +8601,8 @@ def cmd_gui(args: argparse.Namespace):
             launch_command.append("--no-sandbox")
         else:
             sys.exit(1)
+    elif _desktop_linux_needs_disable_setuid_sandbox(packaged_executable):
+        launch_command.append("--disable-setuid-sandbox")
 
     launch_command.extend(config_electron_flags)
     print(f"→ Launching packaged Hermes Desktop: {' '.join(launch_command)}")
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -8098,6 +8098,37 @@ def _desktop_linux_needs_no_sandbox() -> bool:
         return False
 
 
+def _desktop_linux_userns_sandbox_available() -> bool:
+    """Return True when Chromium's unprivileged user-namespace sandbox works.
+
+    When an unprivileged process can create a user namespace, Chromium uses the
+    namespace sandbox and never consults the setuid ``chrome-sandbox`` helper,
+    so requiring the helper to be root-owned 4755 (and prompting for sudo) is
+    unnecessary. Probe the real capability with ``unshare`` instead of reading
+    distro-specific sysctls: the probe fails closed on hosts where user
+    namespaces are disabled or AppArmor-restricted, which then follow the
+    existing setuid-helper path.
+    """
+    if sys.platform != "linux":
+        return False
+    unshare = shutil.which("unshare")
+    if not unshare:
+        return False
+    try:
+        return (
+            subprocess.run(
+                [unshare, "--user", "--map-root-user", "true"],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                timeout=5,
+                check=False,
+            ).returncode
+            == 0
+        )
+    except (OSError, subprocess.TimeoutExpired):
+        return False
+
+
 def _desktop_linux_sandbox_helper_is_regular_file(packaged_executable: Path) -> bool:
     """Return True when ``chrome-sandbox`` exists as a regular file."""
     if sys.platform != "linux":
@@ -8134,6 +8165,10 @@ def _desktop_linux_sandbox_fixup(packaged_executable: Path) -> bool:
         return False
 
     if sandbox_lstat.st_uid == 0 and stat.S_IMODE(sandbox_lstat.st_mode) == 0o4755:
+        return True
+
+    if _desktop_linux_userns_sandbox_available():
+        print("✓ Using Chromium's user-namespace sandbox (setuid helper not needed).")
         return True
 
     sudo = shutil.which("sudo")
--- a/scripts/desktop-update/posix.sh
+++ b/scripts/desktop-update/posix.sh
@@ -316,6 +316,10 @@ linux_gate() {
   sb="$unpacked/chrome-sandbox"
   if [ ! -e "$sb" ]; then GATE=relaunch; return; fi
   if [ -u "$sb" ] && [ "$(stat -c %u "$sb" 2>/dev/null)" = "0" ]; then GATE=relaunch; return; fi
+  # Namespace sandbox usable => Electron never consults the setuid helper,
+  # so a non-root chrome-sandbox does not block relaunch (mirrors the
+  # _desktop_linux_userns_sandbox_available() probe in hermes_cli/main.py).
+  if unshare --user --map-root-user true 2>/dev/null; then GATE=relaunch; return; fi
 
   case "${ELECTRON_DISABLE_SANDBOX:-}" in 1|true|TRUE|True) GATE=relaunch; return ;; esac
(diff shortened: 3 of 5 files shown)


Maxxim69:
Reproduced again today (2026-08-30) on a second machine of the same class
as the live-host test in the PR description (CachyOS,
`kernel.unprivileged_userns_clone=1`,
`/proc/sys/kernel/apparmor_restrict_unprivileged_userns` absent):

- `hermes update` (v0.20.6) repacked the desktop app and reset
  `apps/desktop/release/linux-unpacked/chrome-sandbox` to user-owned
  `0755`. Menu launch (`Terminal=false`) then dies silently: no TTY for
  the sudo prompt, and because the sysctl file is absent the
  `--no-sandbox` fallback never triggers (same gap #96687 addresses).
- This PR's probe passes on the machine:
  `unshare --user --map-root-user true` → exit 0, so both (1) the
  sudo-skip in `_desktop_linux_sandbox_fixup` and (2)
  `--disable-setuid-sandbox` should work here without sudo.
- One triage-relevant data point: the "passwordless sudo configured for
  exactly those two commands" workaround mentioned in #51327 is not
  achievable in sudoers syntax — `root:root` as a command argument is a
  parse error (rejected by `visudo -c` in both inline and `Cmnd_Alias`
  form), and the only `chown` spec that parses is a wildcard, which is
  too broad to grant. So on unrestricted-userns hosts the setuid-helper
  path is effectively unrepairable by the launcher, making this userns
  probe the only no-sudo route. Happy to share the exact visudo test
  output if useful.

Workaround in place meanwhile: one-time `chown root:root` +
`chmod 4755` on the helper (via pkexec/polkit, which works where sudo
cannot prompt).

Maxxim69:
Reproduced again (2026-08-30) on a second machine of the same class as the PR's live-host test (CachyOS, `kernel.unprivileged_userns_clone=1`, `/proc/sys/kernel/apparmor_restrict_unprivileged_userns` absent):

- `hermes update` (v0.20.6) repacked the desktop app and reset `apps/desktop/release/linux-unpacked/chrome-sandbox` to user-owned `0755`. The `.desktop` entry (`Terminal=false`) then exits silently: no TTY for the sudo prompt, and because the AppArmor sysctl file is absent the `--no-sandbox` fallback never triggers.
- This PR's probe passes on the box (`unshare --user --map-root-user true` → exit 0), so both (1) the sudo-skip in `_desktop_linux_sandbox_fixup` and (2) `--disable-setuid-sandbox` should apply without sudo.
- One data point that closes off the alternative workaround from #51327 ("configure passwordless sudo for exactly those two commands"): it isn't expressible in sudoers. `root:root` as a command argument is a hard parse error — `visudo -c` rejects it in both inline and `Cmnd_Alias` form — and the only `chown` spec that parses is a wildcard, which is too broad to grant. So on unrestricted-userns hosts the setuid-helper path is effectively unrepairable by the launcher, which is why the fail-closed userns probe is the right no-sudo route.

The added `linux_gate()` probe in `scripts/desktop-update/posix.sh` is the only change that touches the update pipeline, so it's worth checking against the updater design rule in `AGENTS.md`: "Every stage exists because its absence was a real field failure; PRs that weaken a stage need to answer for the failure class it guards." My read is that this *strengthens* rather than weakens the gate. The gate's failure class is a relaunch that "quit and never came back" because Electron aborts on a present-but-non-setuid helper; the probe only unlocks the relaunch when `unshare --user --map-root-user` succeeds, i.e. when Chromium uses the namespace sandbox and (per the PR's premise) never consults the setuid helper, so that failure class cannot occur on the host. On hosts where the probe fails (userns disabled or AppArmor-restricted) the existing `ELECTRON_DISABLE_SANDBOX` / `--sandbox-fallback` / `--no-sandbox` / manual path is untouched.

Workaround in place meanwhile: a one-time `chown root:root` + `chmod 4755` on the helper (via pkexec/polkit, which can prompt where sudo cannot).

SemihMutlu07:
Found one remaining update-handoff seam while validating this fix.

`linux_gate()` now returns `relaunch` when the userns probe succeeds, but `launch_app()` replays `RELAUNCH_ARGS` unchanged. If the app started before the rebuild with a working `root:root 4755` helper (or from a path that did not already carry the new switch), the rebuild can leave a present user-owned `0755` helper and relaunch without `--disable-setuid-sandbox`. That bypasses the safeguard added by this PR's second commit for Electron builds that reject the present-but-unconfigured helper.

I opened #99561 directly against the maintainer-prepared `fix/p1-chrome-sandbox` branch. It appends the switch exactly once after the same fail-closed userns probe and adds a full hand-off regression that verifies the relaunched process receives both the sandbox switch and its original deep-link arg.

Verified locally: gate matrix green, launch matrix green, focused Python suite 52 passed / 9 skipped.

Prepared with GPT-5.6-sol via pi; commands and outputs were verified locally.

teknium1:
Thanks @kvnloo — landed on main via #100136 with authorship preserved (userns namespace sandbox when the kernel allows; never --no-sandbox; sudo path kept as automatic fallback). Closing as superseded by the merged salvage.

PR #98137, opened 2026-08-29
fix(compression): harden lease refresh and failure backoff

vsd2807:
## What does this PR do?

Hardens the context-compression lease and failure-retry paths for large sessions.

This PR addresses two failure modes identified in [#97948](https://github.com/NousResearch/hermes-agent/issues/97948):

1. A long-running compression could outlive the lease refresher's retry window after transient database contention. When the worker eventually reached publication, the lease could be lost and the completed compression discarded.
2. A failed compression split/publication could immediately be retried on the next turn, causing the same expensive compression to be attempted repeatedly without giving the failure path time to recover.

The changes build on the existing compression and retry mechanisms rather than introducing a separate compression path.

## Related Issue

Refs [#97948](https://github.com/NousResearch/hermes-agent/issues/97948)

## Type of Change

### Lease refresh and ownership handling

- Distinguish successful lease renewal, transient database/SQLite failures, and genuine ownership loss.
- Keep the compression lease refresher alive through transient refresh failures instead of permanently giving up after the previous failure threshold.
- Stop the refresher immediately when the lock is genuinely lost or has been acquired by another holder.
- Preserve the holder-qualified lease semantics so an expired lease can only be revived by its existing holder.
- Keep the pre-publication lease refresh, ownership check, and publication protected by the same database transaction.
- Preserve the existing behavior for callers that do not request the additional pre-publication refresh.

### Compression failure backoff

- Record `session_split_failed` through the existing durable compression failure/backoff mechanism.
- Use the existing backoff ladder rather than introducing a second independent cooldown.
- Prevent an automatic retry from immediately repeating the same failed compression generation.
- Preserve the existing forced/manual compression bypass.

### Tests

Added and updated regression coverage for:

- successful lease renewal
- transient SQLite/write failures
- persistent transient failures
- recovery after transient failures
- genuine ownership loss
- expired-but-still-owned leases
- concurrent holder replacement
- atomic pre-publication lease refresh
- split/publication failure backoff
- cooldown expiry
- forced compression bypass
- compatibility with existing compression/lease callers

The timeout/terminal-settlement problem described in Symptom A of [#97948](https://github.com/NousResearch/hermes-agent/issues/97948) is intentionally **not addressed by this PR**. The earlier timeout-reconciliation approach was removed after review because a correct fix requires durable attempt identity, late-settlement handling, and canonical session/generation projection rather than a one-shot database lookup after the gateway timeout. That work is better handled separately.

The lean-tail performance regression described as Symptom C in the issue has also been addressed separately upstream.

### 1. Lease refresher behavior

```bash
pytest tests/state/test_compression_lease_typed_refresh.py
```

These tests verify that transient refresh failures do not terminate the refresher prematurely, including failures beyond the previous threshold, while genuine ownership loss still stops it immediately.

### 2. Pre-publication lease refresh

```bash
pytest tests/state/test_compression_lease_refresh_before_publish.py
```

This verifies that the final lease refresh and ownership validation remain atomic and that a stale holder cannot extend or publish against a lock that has been reclaimed by another holder.

### 3. Split failure backoff

```bash
pytest tests/agent/test_compression_split_failed_backoff.py
```

This verifies that `session_split_failed` records the existing durable compression backoff and that forced/manual compression can still bypass it.

### 4. Existing compression/lease regression coverage

```bash
pytest tests/test_hermes_state.py -k compression
pytest tests/state/test_compression_lineage_guard.py
pytest tests/agent/test_compression_concurrent_fork.py
```

The relevant existing suites should be run alongside the new regression tests before merging.

### DIFF
--- a/agent/conversation_compression.py
+++ b/agent/conversation_compression.py
@@ -2277,11 +2277,12 @@ class _CompressionLockLeaseRefresher:
         if refresh_interval_seconds is None:
             refresh_interval_seconds = max(1.0, min(60.0, ttl_seconds / 2.0))
         self._refresh_interval_seconds = max(0.1, float(refresh_interval_seconds))
-        # Tolerate transient refresh failures for at most one lease's worth of
-        # time, so the give-up window is genuinely bounded by the TTL the
-        # acquirer set (a single blip recovers on the next tick; a persistent
-        # failure stops before the lease could outlive its TTL). Floor of 1 so a
-        # degenerate interval >= ttl still tolerates one blip.
+        # Transient blips are tolerated indefinitely while the holder still owns
+        # the row (holder-only semantics guarantee an expired-but-unreclaimed
+        # row revives as ``renewed``). ``_max_consecutive_failures`` is the
+        # interval-bounded log threshold (ttl/interval, e.g. 5 for 300/60) —
+        # reaching it logs "continuing to probe" but does not stop the loop.
+        # Only a proven ``ownership_lost`` stops the thread.
         self._max_consecutive_failures = max(
             1, int(self._ttl_seconds / self._refresh_interval_seconds)
         )
@@ -2309,13 +2310,15 @@ class _CompressionLockLeaseRefresher:
     def _run(self) -> None:
         # A single falsy refresh must NOT permanently kill the lease: a
         # transient DB blip (write contention escaping _execute_write's retry
-        # budget, a momentary "database is locked") returns False just like a
-        # genuine lost-ownership, but only the latter should stop the loop.
-        # Tolerate consecutive failures for at most one lease's worth of time
-        # (_max_consecutive_failures = ttl / interval), so a one-off blip
-        # recovers on the next tick while the total give-up window stays bounded
-        # by the TTL the acquirer set — the lock can never be held past its TTL
-        # by a stuck refresher.
+        # budget, a momentary "database is locked") returns ``transient_failure``
+        # just like a genuine lost-ownership ``ownership_lost``, but only the
+        # latter should stop the loop.
+        # ``_max_consecutive_failures`` still caps the consecutive transient
+        # window: a one-off blip recovers on the next tick, but a persistent
+        # transient burst is logged and probed with bounded backoff. Only a
+        # proven ``ownership_lost`` stops the thread immediately — holder-only
+        # semantics guarantee a revivable expired-but-unreclaimed row returns
+        # ``renewed`` even past its TTL.
         consecutive_failures = 0
         # First refresh happens immediately, not one interval late. Everything
         # between try_acquire() and start() (the rotation-ownership lookup, the
@@ -2336,18 +2339,42 @@ class _CompressionLockLeaseRefresher:
                 )
             except Exception as exc:
                 logger.debug("compression lock refresh raised: %s", exc)
-                refreshed = False
-            if refreshed:
+                refreshed = "transient_failure"
+            # Backward-compat: old mocks / callers may return bool
+            if refreshed is True or refreshed == "renewed":
                 consecutive_failures = 0
                 continue
-            consecutive_failures += 1
-            if consecutive_failures >= self._max_consecutive_failures:
+            if refreshed == "ownership_lost":
                 logger.debug(
-                    "compression lock refresh failed %d times in a row; "
-                    "stopping lease refresher for session %s",
-                    consecutive_failures, self._session_id,
+                    "compression lock ownership lost for session %s (holder %s); "
+                    "stopping lease refresher",
+                    self._session_id,
+                    self._holder,
                 )
                 break
+            # ``transient_failure`` (or legacy ``False``/``None``)
+            if refreshed == "transient_failure" or refreshed is False or refreshed is None:
+                consecutive_failures += 1
+                if consecutive_failures >= self._max_consecutive_failures:
+                    logger.debug(
+                        "compression lock refresh transient %d times in a row; "
+                        "continuing to probe (holder %s still owns row) for session %s",
+                        consecutive_failures,
+                        self._holder,
+                        self._session_id,
+                    )
+                # Do NOT break — keep probing while holder still owns the row.
+                # The interval wait at the top of the loop bounds the retry.
+                continue
+            # Unknown string — treat as transient to avoid silently holding a
+            # stolen lease, but log and continue.
+            logger.debug(
+                "compression lock refresh unknown result %r for session %s; treating as transient",
+                refreshed,
+                self._session_id,
+            )
+            consecutive_failures += 1
+            continue
 
 
 def check_compression_model_feasibility(agent: Any) -> None:
@@ -4714,6 +4741,8 @@ def compress_context(
                         profile_name=_profile_for_child,
                         compression_lock_holder=_lock_holder,
                         require_compression_lease=_lock_holder is not None,
+                        require_lease_refresh=_lock_holder is not None,
+                        lease_ttl_seconds=_lock_ttl,
                         watermark=(
                             _commit_watermark
                             if _foreign_tail_ceiling is not None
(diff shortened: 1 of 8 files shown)


Enough1122:

**Overall:** Fixes three related compression failure modes that leave large sessions stuck or waste work: gateway timeout after successful commit, transient lease-refresher failure, and repeated split-failure retries. Well-bundled, heavily tested.

**What it does**
- **RC1 — timeout reconciliation:** `tui_gateway/methods_session.py:session.compress` catches `Exception` around `_send_compute_host_control(timeout=120)`. On timeout, calls `_reconcile_compression_after_timeout(parent_id)` which checks `find_live_compression_child` (rotation) → child messages, or `get_session(...).end_reason=="compression"` (in-place) — returns `{"status":"compressed","turn_isolation":True,...}` on success, `None` otherwise. On `None`, returns original `5019` error; exceptions in reconciliation are swallowed to `None` so original error is not obscured (tests pin this). Child messages, not parent, are returned for rotation.
- **RC2 — lease refresh:** `hermes_state.py:publish_compression_child` gains `require_lease_refresh` + `lease_ttl_seconds`. When true, does `UPDATE compression_locks SET expires_at = now+ttl WHERE session_id=? AND holder=?` *inside same transaction* before expiry check — gives a transient-DB-failure refresher one final chance, no TOCTOU. Same `conn` as publication. Tests pin atomicity via intercepted `_execute_write`.
- **RC3 — split-failure cooldown:** `agent/conversation_compression.py` now `_record_compression_failure_cooldown(60, f"session_split_failed: {e}")` when `publish_compression_child` / indexing fails, so automatic compression is locally blocked for 60s (manual `force=True` bypasses, verified).
- **Desktop:** `SESSION_COMPRESS_TIMEOUT_MS` 120000 → 150000 to not time out before gateway's 120s reconciliation.
- Tests: `test_compress_timeout_reconciliation.py` (rotation, inplace, still-running, abort, DB-unavailable, RPC integration, race-boundary rejects, exception-does-not-obscure, child-vs-parent messages) and `test_compression_lease_refresh_before_publish.py` (5 lease cases) + cooldown tests.

**Non-blocking notes**
- `_reconcile_compression_after_timeout` trusts `end_reason=="compression"` for in-place — ensure no other path sets that reason erroneously (only compression should). Rejects unrelated `end_reason` like `reset`/`archive` correctly (tested).
- `require_lease_refresh` does `UPDATE` before `SELECT holder,expires_at` — if holder mismatched, update affects 0 rows, then expiry check correctly raises `CompressionSessionBusyError("lease lost")` (tested).
- `get_messages_as_conversation` for child could be expensive for huge sessions; acceptable for rare timeout path, but consider limiting if sessions are massive.
- Desktop 150s vs gateway 120s leaves 30s margin — reasonable; ensure desktop's overall WS request timeout also covers this.

Comprehensive fix for production large-session pain.

andrexibiza (commented):
The in-transaction pre-publication lease refresh is a useful narrow repair. The timeout reconciliation is not a valid fix for #97948 yet.

### Blocker — RC1 queries `state.db` with the runtime UI id, not the persisted session id

`session.create` deliberately allocates two identities: the RPC/runtime `sid` and a separate durable `key`, then stores the latter in `session["session_key"]`:

```py
sid = uuid.uuid4().hex[:8]
key = _new_session_key()
...
"session_key": key,
```

This branch calls:

```py
reconciled = _reconcile_compression_after_timeout(sid)
```

and `_reconcile_compression_after_timeout()` uses that value directly in `find_live_compression_child()` / `get_session()`. Those APIs address persisted session rows, so production reconciliation looks for the runtime `sid` instead of `session["session_key"]` / `agent.session_id`.

The tests encode the same mismatch rather than exposing it: `_make_session()` carries `session_key="sess-compress"`, while `test_session_compress_rpc_reconciles_after_timeout` creates `sid="sid-timeout-reconcile"` and seeds the mock DB under that `sid`. A real-DB integration test with distinct runtime and stored ids should fail this branch.

### Blocker — the one-shot check still reports terminal failure for the reported late commit

The check runs exactly once, immediately after the 120-second waiter expires. If the worker is still running, the helper returns `None` and the handler immediately returns `5019`; the worker retains commit authority and its later ack is still dropped. The issue reports completion around 6.4 minutes and another attempt at `total_duration_ms=710109`, so neither can be observed by a one-shot read at 120 seconds. Raising the Desktop budget to 150 seconds does not change this—the gateway has already returned the false terminal error at roughly 120 seconds.

`test_session_compress_rpc_reconciles_after_timeout` pre-seeds the child before the timeout is raised. It covers only “commit already visible when the waiter fails,” not “waiter expires, worker commits minutes later,” which is the actual defect.

### Blocker — even a detected rotation is not projected into canonical runtime identity

The normal compute-host terminal ack carries the rotated `session_key`, `history_version`, `message_count`, and `session_info`. The fallback returns `host_ack: {}` plus messages, does not return the canonical child/session id, and does not update the gateway session shadow. That can present child transcript content while subsequent controls remain bound to the stale parent identity—the exact projection divergence #97948 requires us to close.

This path also catches every `Exception`, not only a timeout, and reconciliation is not attempt-, generation-, or watermark-scoped. An unrelated control failure can therefore be converted into success from pre-existing lineage state.

### Required proof before this can own #97948

- Use the persisted session key, with a real test proving runtime `sid != session_key`.
- Let a worker cross the control deadline and commit later; preserve/query that exact attempt and reconcile the late terminal result.
- Apply the returned canonical child identity and generation to the gateway/Desktop projection, with stale-completion rejection.
- Bind success to an attempt id or input generation, not merely “some compression child exists.”
- Rebase onto current main and reacquire exact-head CI/Docker/Nix; the current head’s workflow runs are all `action_required`.

Until then, `Fixes #97948` overclaims. The lease-refresh hunk can remain a separate narrow owner, but RC1 does not close symptom A.

vsd2807:
Thanks for taking the time to trace this @andrexibiza , all the way through the gateway/compute-host boundary. I went back through the issue and the related PRs with your comments in mind, and I agree that the original RC1 was too narrow to claim the terminal-settlement problem was fixed.

In particular, the distinction between the runtime sid and persisted session_key, the one-shot reconciliation window, and the missing canonical projection made it clear that trying to patch the timeout path here would leave the underlying settlement problem unresolved.

I've therefore narrowed this PR rather than trying to stretch the fix further:

RC2: distinguish transient lease-refresh failures from actual ownership loss, so a temporary DB/write-contention failure doesn't permanently kill a refresher that still owns the lease. Genuine ownership loss still terminates the refresher, and the final pre-publication refresh remains holder-qualified and atomic.
RC3: feed session_split_failed into the existing durable compression backoff mechanism instead of maintaining a separate fixed cooldown, so we don't immediately repeat the same expensive failed compression.

I also added regression coverage for the refresher state machine, including transient failures that persist beyond the old failure threshold and genuine ownership loss, along with the publication interlock.

The original RC1 timeout reconciliation and Desktop timeout increase have been removed from this PR. I think the attempt/settlement/generation work needs to be designed as a separate change rather than having this PR claim ownership of it prematurely.

I've also rebased the branch onto current main and verified the narrowed changes and tests against that base.

Thanks again for the detailed review — it definitely helped make the scope and failure boundaries much clearer.

kshitijk4poor:
Thanks @vsd2807 — the two solid pieces of this PR are landing. #99216 salvages them onto current `main` with your authorship preserved on the production commit (cherry-picked content, author `VVV`), and carries `Closes #98137` so this closes on merge.

**What was carried:**
1. The in-transaction pre-publication lease refresh (`require_lease_refresh` / `lease_ttl_seconds` on `publish_compression_child`) — exactly the piece @andrexibiza's review called "a useful narrow repair" that "can remain a separate narrow owner". Your 6 lease tests ride along (re-homed to `tests/state/`), including both wrong-holder cases.
2. The split-failure cooldown — plus your 3 unit tests, of which the force-bypass one was kept verbatim; two were dropped as duplicates of existing coverage in `test_compression_rotation_state.py` / `test_context_compressor.py` (stated in the salvage's test docstring). Follow-up commits guard the recording with the sibling try/except convention, name the constant (`_SPLIT_FAILURE_COOLDOWN_SECONDS`), and add an E2E test through a real `SessionDB`.

**What was NOT carried, and why:** the timeout-reconciliation half (tui_gateway/server.py, methods_session.py, the desktop 150s bump, and its 490-line test file). @andrexibiza's review identified three structural blockers there (runtime `sid` vs persisted `session_key`, the one-shot check being unable to observe a multi-minute commit, and no identity projection of a detected rotation) — that half needs the redesign his review outlines, and re-attempting it inside a salvage would not have done it justice. #97948 therefore stays open for symptom A (the salvage says `Refs`, not `Fixes`).

If you take another run at the reconciliation half against his required-proof list, that would be a genuinely valuable PR — the lease groundwork will already be on main.

PR #98140, opened 2026-08-29
fix(mcp): let Railway OAuth login reach authorization

### LINKED ISSUE #98117: [Bug]: MCP OAuth authorization URL uses '?' instead of '&' when authorization_endpoint already has a query string — breaks Railway login (missing response_type)
### Summary

`hermes mcp login <server>` builds a malformed authorization URL when the provider's `authorization_endpoint` already contains a query string. The next parameter is appended with `?` instead of `&`, so the first param (`response_type`) is swallowed into the preceding value and the provider rejects the request.

Reproduced against the **Railway** MCP server (`https://mcp.railway.com`), which is in the Hermes catalog (`hermes mcp install railway`).

### Environment

- Hermes Agent v0.20.6 (2026.8.27) · upstream 4209d371
- mcp SDK 2.0.0
- macOS 26.5.2
- Install method: git

### Root cause

Railway's discovery document returns an `authorization_endpoint` that **already has a query string**:

```console
$ curl -s https://mcp.railway.com/.well-known/oauth-authorization-server
{"issuer":"https://backboard.railway.com",
 "authorization_endpoint":"https://backboard.railway.com/oauth/auth?resource=https%3A%2F%2Fmcp.railway.com",
 "token_endpoint":"https://backboard.railway.com/oauth/token",
 ...}
```

The URL is then assembled with an unconditional `?` — `venv/lib/python3.11/site-packages/mcp/client/auth/oauth2.py:415`:

```python
authorization_url = f"{auth_endpoint}?{urlencode(auth_params)}"
```

When `auth_endpoint` already contains `?`, the result has two of them:

```
https://backboard.railway.com/oauth/auth?resource=https%3A%2F%2Fbackboard.railway.com?response_type=code&client_id=...
                                                                                      ^ should be &
```

Per RFC 3986 everything after the first `?` is the query string, so `resource` parses as the literal value `https://backboard.railway.com?response_type=code` and `response_type` never exists as a parameter.

### Steps to reproduce

1. `hermes mcp install railway`
2. `hermes mcp login railway`
3. Open the printed URL in a browser.

### Actual behavior

The provider redirects to the callback with:

```
error=invalid_request&error_description=missing+required+parameter+%27response_type%27
```

Authentication can never complete. Every attempt fails identically.

### Expected behavior

The authorization URL should be built with the correct separator, joining with `&` when the endpoint already carries a query string.

### Workaround (confirmed)

Hand-editing the printed URL to replace the second `?` with `&` produces a working authorization request that reaches Railway's consent screen and returns a valid code.

### Suggested fix

Choose the separator based on whether the endpoint already has a query, rather than always using `?`:

```python
sep = "&" if urlparse(auth_endpoint).query else "?"
authorization_url = f"{auth_endpoint}{sep}{urlencode(auth_params)}"
```

`urlparse` is already imported at `oauth2.py:15`.

This is upstream SDK code, but it makes a catalog-listed MCP server (`railway`) unusable through `hermes mcp login`, so it may warrant a local fix or a pinned patch until the SDK is updated. Note that RFC 8707 also has the flow set `resource` itself, so the merged URL ends up with two `resource` params — worth deciding which should win.

### Related

Possibly relevant to #96107 (monday.com AS metadata mismatch) — both are cases where a provider's discovery document doesn't match the SDK's assumptions.

fangliquanflq:
## What does this PR do?

Railway MCP login now reaches the authorization page when discovery returns an authorization endpoint that already contains query parameters. Hermes preserves the endpoint query and fragment, appends the MCP SDK's OAuth parameters with the correct separator, and avoids repeating an identical RFC 8707 resource value.

### Symptom

`hermes mcp login railway` prints an authorization URL with a second `?`. Railway parses `response_type=code` as part of the existing `resource` value and rejects login with `missing required parameter 'response_type'`.

### Impact

OAuth login cannot complete for catalog MCP servers whose discovered authorization endpoint already contains a query string, including Railway.

### Bug Cause

**Trigger:** `mcp.client.auth.oauth2.OAuthClientProvider._perform_authorization_code_grant` when `authorization_endpoint` already has a query component.

**Causal chain:**
1. OAuth discovery returns an authorization endpoint containing `resource=...`.
2. MCP SDK 2.0.0 unconditionally concatenates `?` plus its generated authorization parameters.
3. The resulting second `?` leaves `response_type` inside the first resource value, so the authorization server cannot find the required parameter.

**Why it is wrong:** RFC 3986 permits only one query delimiter. Client parameters must be joined to an existing query with `&` and must appear before any fragment.

**Working sibling / contrast:** Endpoints without a query string already work because the SDK's unconditional `?` is valid there.

**Ruled out:** OAuth discovery and callback handling are not the failure source. The captured production-provider URL shows the generated parameters are correct but placed behind the wrong delimiter.

### Fix

A shared compatibility wrapper now repairs the SDK-generated URL in both Hermes OAuth provider paths. It preserves the endpoint's raw query encoding and fragment, removes only an identical generated `resource` value, and retains distinct RFC 8707 resource parameters.

## Related Issue

Fixes #98117

## Type of Change

## Changes Made

- `tools/mcp_oauth.py` - merge generated OAuth parameters into discovered authorization endpoint queries safely.
- `tools/mcp_oauth_manager.py` - apply the shared compatibility wrapper to the production managed provider.
- `tests/tools/test_mcp_oauth_authorization_url.py` - cover existing queries, fragments, encoding, duplicate and distinct resource values, and query-free endpoints.

## How to Test

1. Run the focused MCP OAuth regression and integration tests:

```bash
scripts/run_tests.sh tests/tools/test_mcp_oauth_authorization_url.py tests/tools/test_mcp_oauth.py tests/tools/test_mcp_oauth_manager.py tests/tools/test_mcp_oauth_metadata.py tests/tools/test_mcp_oauth_integration.py -q
```

2. Confirm all 82 tests pass with one existing skip.
3. Run Ruff on the changed Python files and `git diff --check`.

### Code

- [x] I've updated relevant documentation (README, `docs/`, docstrings) - N/A; no user-facing configuration changed

## Screenshots / Logs

N/A - the regression test captures and validates the production provider's authorization URL directly.

### DIFF
--- a/tests/tools/test_mcp_oauth_authorization_url.py
+++ b/tests/tools/test_mcp_oauth_authorization_url.py
@@ -1,0 +1,120 @@
+"""Authorization URL compatibility tests for MCP OAuth providers."""
+
+from urllib.parse import parse_qs, urlsplit
+
+import pytest
+from mcp.shared.auth import (
+    OAuthClientInformationFull,
+    OAuthClientMetadata,
+    OAuthMetadata,
+)
+from pydantic import AnyUrl
+
+from tools.mcp_oauth import HermesTokenStorage
+from tools.mcp_oauth_manager import _HERMES_PROVIDER_CLS
+
+
+class _AuthorizationURLCaptured(Exception):
+    pass
+
+
+async def _capture_authorization_url(tmp_path, authorization_endpoint: str) -> str:
+    captured: list[str] = []
+
+    async def redirect_handler(url: str) -> None:
+        captured.append(url)
+        raise _AuthorizationURLCaptured
+
+    async def callback_handler():  # pragma: no cover - redirect stops first
+        raise AssertionError("callback must not run")
+
+    provider = _HERMES_PROVIDER_CLS(
+        server_name="railway",
+        server_url="https://mcp.railway.com",
+        client_metadata=OAuthClientMetadata(
+            redirect_uris=[AnyUrl("http://127.0.0.1:12345/callback")],
+            client_name="Hermes Agent",
+        ),
+        storage=HermesTokenStorage("railway", hermes_home=tmp_path),
+        redirect_handler=redirect_handler,
+        callback_handler=callback_handler,
+    )
+    provider.context.client_info = OAuthClientInformationFull(
+        client_id="test-client",
+        redirect_uris=[AnyUrl("http://127.0.0.1:12345/callback")],
+        grant_types=["authorization_code", "refresh_token"],
+        response_types=["code"],
+        token_endpoint_auth_method="none",
+    )
+    provider.context.oauth_metadata = OAuthMetadata.model_validate(
+        {
+            "issuer": "https://backboard.railway.com",
+            "authorization_endpoint": authorization_endpoint,
+            "token_endpoint": "https://backboard.railway.com/oauth/token",
+            "response_types_supported": ["code"],
+        }
+    )
+    provider.context.protocol_version = "2025-06-18"
+
+    with pytest.raises(_AuthorizationURLCaptured):
+        await provider._perform_authorization_code_grant()
+
+    return captured[0]
+
+
+@pytest.mark.asyncio
+async def test_authorization_url_joins_existing_query_and_deduplicates_resource(
+    tmp_path,
+):
+    url = await _capture_authorization_url(
+        tmp_path,
+        "https://backboard.railway.com/oauth/auth?"
+        "resource=https%3A%2F%2Fmcp.railway.com",
+    )
+
+    parsed = urlsplit(url)
+    params = parse_qs(parsed.query)
+    assert params["response_type"] == ["code"]
+    assert params["resource"] == ["https://mcp.railway.com"]
+    assert parsed.fragment == ""
+
+
+@pytest.mark.asyncio
+async def test_authorization_url_preserves_existing_encoding_and_fragment(tmp_path):
+    url = await _capture_authorization_url(
+        tmp_path,
+        "https://backboard.railway.com/oauth/auth?"
+        "audience=hello%20world&opaque=a%2Bb#consent",
+    )
+
+    parsed = urlsplit(url)
+    assert parsed.query.startswith("audience=hello%20world&opaque=a%2Bb&")
+    assert parse_qs(parsed.query)["response_type"] == ["code"]
+    assert parsed.fragment == "consent"
+
+
+@pytest.mark.asyncio
+async def test_authorization_url_preserves_distinct_resource_parameters(tmp_path):
+    url = await _capture_authorization_url(
+        tmp_path,
+        "https://backboard.railway.com/oauth/auth?"
+        "resource=https%3A%2F%2Fapi.railway.com",
+    )
+
+    assert parse_qs(urlsplit(url).query)["resource"] == [
+        "https://api.railway.com",
+        "https://mcp.railway.com",
+    ]
+
+
+@pytest.mark.asyncio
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Fixes Railway (and similar) OAuth where discovery endpoint already contains a query (`?resource=...`). SDK appended a second `?`, corrupting `response_type=code` into query fragment and breaking authorization.

**What it does**
- `tools/mcp_oauth.py` adds `_join_authorization_endpoint_query(authorization_endpoint, authorization_url)` — detects `authorization_url.startswith(endpoint+"?")`, splits endpoint via `urlsplit`, extracts existing `resource` values, filters generated `resource` duplicates (`value not in existing_resources`), `urlencode`s remaining generated params, merges (`endpoint.query + "&" + encoded_generated`), preserves `fragment` via `urlunsplit`. Identical RFC 8707 `resource` wins; distinct values remain repeated.
- Wraps grant via `_perform_authorization_code_grant_with_compatible_query(provider, parent_method)` that swaps `redirect_handler` to repair URL, delegates to `super()._perform...`, restores handler in `finally` (identity check).
- Patched in both provider classes: `HermesOAuthClientProvider` (legacy `tools/mcp_oauth.py`) and `HermesMCPOAuthProvider` (`tools/mcp_oauth_manager.py`, dynamic).
- Tests `test_mcp_oauth_authorization_url.py` capture URL via `redirect_handler` injection: joins existing query & dedup, preserves encoding + fragment (`audience=...%20world&opaque=a%2Bb#consent`), distinct resources both present, no-query unchanged.

**Non-blocking notes**
- Query merge preserves existing encoding (`hello%20world`, `a%2Bb`) by using `parse_qsl` then `urlencode` — `urlencode` will re-encode as `+` vs `%20` potentially; test asserts `startswith("audience=hello%20world&opaque=a%2Bb&")` — note `urlencode` encodes space as `+` by default; here original used `%20`, but `parse_qsl` decodes then re-encodes — verify SDK's original `%20` is preserved as `%20` vs converted to `+`. Test expects `%20` preserved, so check implementation uses `quote_via` or that `urlencode` with `quote` handles `%20` vs `+`; test passing confirms compatibility.
- Fallback `return authorization_url` when prefix mismatch is safe — non-Railway providers without query pass through unchanged.
- Restores handler only if `provider.context.redirect_handler is redirect_handler` — guards against handler replaced mid-grant by concurrent code.

Correct interop fix.

fangliquanflq:
Thanks for the review. I checked each note against the committed implementation:

- The endpoint's existing query is preserved byte-for-byte: `_join_authorization_endpoint_query` reads `endpoint.query` only to compare decoded `resource` values, then appends generated parameters to the original raw `endpoint.query`. It does not re-encode `audience=hello%20world&opaque=a%2Bb`; only the SDK-generated suffix passes through `urlencode`.
- The prefix-mismatch path intentionally leaves providers that do not exhibit the MCP SDK 2.0 URL shape unchanged.
- The identity check in `finally` restores the original handler when this wrapper still owns the slot, while avoiding overwriting a handler replacement made during the grant.

No code change was needed. I verified the focused encoding/fragment assertion and the surrounding OAuth paths with:

`scripts/run_tests.sh tests/tools/test_mcp_oauth_authorization_url.py tests/tools/test_mcp_oauth.py tests/tools/test_mcp_oauth_manager.py tests/tools/test_mcp_oauth_metadata.py tests/tools/test_mcp_oauth_integration.py -q`

Result: 82 passed, 1 skipped. Ruff and `git diff --check` also pass.

PR #98141, opened 2026-08-29
fix(mcp): prevent OAuth state rotation during interactive login

### LINKED ISSUE #98118: [Bug]: hermes mcp login rotates OAuth state while the paste prompt is open — 'State parameter mismatch' makes the documented SSH fallback unusable
### Summary

Over SSH, `hermes mcp login <server>` prints an authorization URL and offers "paste the redirect URL here" as the recommended fallback. But the flow re-prompts on its own schedule, and **each re-prompt mints a new `state` and `code_challenge`**. The callback the user pastes is validated against the *newest* state, so a valid callback from the URL they were actually given is rejected as a mismatch.

The documented fallback is effectively unusable over SSH: the user is racing a timer they can't see and can't win.

### Environment

- Hermes Agent v0.20.6 (2026.8.27) · upstream 4209d371
- mcp SDK 2.0.0
- macOS 26.5.2 (remote host: Mac Studio; browser on a separate MacBook over SSH)
- Install method: git

### Steps to reproduce

1. SSH into a remote host running Hermes.
2. `hermes mcp login railway` (any OAuth MCP server reproduces this).
3. Note the printed URL contains `state=A`.
4. Open it in a browser on the local machine, authorize.
5. The redirect to `http://127.0.0.1:<port>/callback` fails (expected — listener is on the remote host).
6. Copy the address-bar URL and paste it at the prompt, as the on-screen instructions direct.

### Actual behavior

```
✗ Authentication failed: State parameter mismatch: A != B
```

where `A` is the state from the URL the user was given, and `B` is a state generated by a later internal re-prompt. Observed across four consecutive attempts with different states each time:

```
✗ Authentication failed: State parameter mismatch:
  LqoLM8q9BHi40JQwfediUvvwp__atX_Bmox5A5CqP8Y != 54lFeOyRtFNXoW4x1LsOg5KJEXe5dpp6t_8q9K0yoDo

✗ Authentication failed: State parameter mismatch:
  xgdv9FJt5R8SXA6gfgK3qlN1KTnPQCMjpH2_OayH83Q != NGv-3AFOcYmzq1vvHQvEohQ_cJS_1WEq36qZ4fOqEvY
```

The process log shows the "MCP OAuth: authorization required / Open this URL" block printed **more than once** in a single `mcp login` invocation, each time with fresh PKCE values — while the earlier prompt is still waiting on stdin.

Notably the SSH **tunnel** path fails the same way: with `ssh -N -L <port>:127.0.0.1:<port>` forwarding correctly, the callback arrives at the listener with the exact state that was issued, and is still rejected because the flow has already rotated.

State validation itself is correct (`oauth2.py:421-422`) — the bug is that the flow rotates state while the user is mid-round-trip.

### Expected behavior

Either:

1. The state/PKCE pair issued with a printed URL stays valid for the full `oauth.timeout` window (no rotation while a paste prompt is open), or
2. If a re-prompt genuinely must issue new parameters, the old ones remain accepted until they expire, or
3. At minimum, the re-prompt prints an explicit warning that the previous URL is now dead, so the user knows to discard it.

Option 1 matches user expectation: one printed URL, one usable authorization.

### Secondary issue: stdin paste is not consumed

Independently, pasting into the process's stdin does not appear to register at all when the login runs under a PTY driven by an agent/automation. The prompt remains displayed unchanged and the value is never processed. Delivering the callback by `curl`-ing the local listener directly (`curl "http://127.0.0.1:<port>/callback?code=...&state=..."`) *does* reach the flow and returns `Authorization Successful` — so the listener works while the stdin reader does not.

### Impact

OAuth MCP setup over SSH is currently only completable by bypassing `hermes mcp login` entirely. I ended up running the PKCE flow manually (generate verifier/challenge, hit `/oauth/auth`, exchange the code at `/oauth/token`, write the token file matching the schema in `~/.hermes/mcp-tokens/<server>.json`) — after which `hermes mcp test railway` reports `✓ Connected (1238ms)` and `✓ Tools discovered: 42`. So the config and token format are fine; only the interactive login path is broken.

Remote/headless installs are a common Hermes deployment, and the on-screen text explicitly advertises the paste fallback as "Easiest", so this affects the documented happy path.

### Related

- #73997 — `mcp login` internal retry self-collides on pinned `oauth.redirect_port`, masking the real auth error. Same "internal retry interferes with the user-facing flow" shape; this issue is the state-rotation consequence.
- #87329 / #80520 (both closed) — headless `mcp login` callback port collisions across retries.

StanleyStetson:
## What does this PR do?

1. **Propagates effective \connect_timeout\ to connection config:** Ensures \_probe_single_server\ forwards the resolved \connect_timeout\ (>= 315.0s during \hermes mcp login\ and re-auth) into the \config\ dictionary passed to \_connect_server\, so underlying HTTP (\streamable_http_client\) and SSE (\sse_client\) transports do not disconnect early at 60s while the user is authenticating in a browser.
2. **Single-flight fail-fast during interactive OAuth:** Updates \MCPServerTask.run\ to check \_is_interactive_forced()\ on initial connection errors and fail fast immediately rather than entering an automatic background reconnection loop that burns retries, rotates PKCE \state\, and mints new authorization URLs while the paste prompt is open.
3. **Cross-SDK result compatibility:** Provides \_LegacyAuthCodeResult\ named tuple fallback for environments where \AuthorizationCodeResult\ is not present in \mcp.shared.auth\.

Fixes #98118

## Why is this change needed?

When running \hermes mcp login <server>\ over SSH or headless environments, the user is instructed to open the authorization URL in a local browser and paste the callback URL into the terminal prompt. If the user takes longer than 60 seconds (the default HTTP transport connect timeout), the underlying HTTP client times out.

Previously, \MCPServerTask.run\ treated this timeout as a transient connection failure and entered an automatic 5-attempt retry loop. Each retry re-ran \_run_http\, minting a new PKCE verifier, challenge, and new \state=B\ while printing duplicate authorization prompts. When the user pasted the callback with \state=A\ from the initial prompt, Hermes rejected it with \State parameter mismatch: A != B\.

## How to Test

- Run affected test suites:
  \\\ash
  scripts/run_tests.sh tests/hermes_cli/test_mcp_config.py tests/tools/test_mcp_oauth.py -q
  \\\
- Verified regression tests:
  - \TestProbeEnvResolution.test_probe_propagates_connect_timeout_to_config\: verifies that \connect_timeout\ passed to \_probe_single_server\ is injected into \config\ passed to \_connect_server\.
  - \	est_interactive_oauth_initial_failure_does_not_retry_loop\: verifies that under \orce_interactive_oauth()\, connection failures fail fast in 1 attempt without looping.

### DIFF
--- a/hermes_cli/mcp_config.py
+++ b/hermes_cli/mcp_config.py
@@ -306,6 +306,17 @@ def _probe_single_server(
             connect_timeout = max(1.0, float(raw_timeout))
         except (TypeError, ValueError):
             connect_timeout = 30.0
+    else:
+        try:
+            connect_timeout = max(1.0, float(connect_timeout))
+        except (TypeError, ValueError):
+            connect_timeout = 30.0
+
+    # Propagate effective connect_timeout into config passed to _connect_server
+    # so that underlying HTTP/SSE/stdio transports honor the full probe timeout
+    # instead of disconnecting early (fixes #98118).
+    config = dict(config)
+    config["connect_timeout"] = connect_timeout
 
     _ensure_mcp_loop()
     tools_found: List[Tuple[str, str]] = []
--- a/tests/hermes_cli/test_mcp_config.py
+++ b/tests/hermes_cli/test_mcp_config.py
@@ -528,6 +528,36 @@ class TestProbeEnvResolution:
         assert tools == [("do_thing", "a tool")]
         assert seen["config"]["headers"]["Authorization"] == "Bearer jwt-token-xyz"
 
+    def test_probe_propagates_connect_timeout_to_config(self, monkeypatch):
+        """_probe_single_server must propagate explicit connect_timeout into the config passed to _connect_server."""
+        import hermes_cli.mcp_config as mc
+
+        seen = {}
+
+        class _FakeTool:
+            name = "do_thing"
+            description = "a tool"
+
+        class _FakeServer:
+            _tools = [_FakeTool()]
+
+            async def shutdown(self):
+                return None
+
+        async def _fake_connect(name, config):
+            seen["config"] = config
+            return _FakeServer()
+
+        monkeypatch.setattr("tools.mcp_tool._connect_server", _fake_connect)
+
+        mc._probe_single_server(
+            "srv",
+            {"url": "http://localhost:5678/mcp", "auth": "oauth"},
+            connect_timeout=315.0,
+        )
+
+        assert seen["config"].get("connect_timeout") == 315.0
+
 
 class TestProbeCapabilityGating:
     """The ``details`` probe must not fire prompts/list or resources/list at
--- a/tests/tools/test_mcp_oauth.py
+++ b/tests/tools/test_mcp_oauth.py
@@ -1112,3 +1112,47 @@ def test_humanize_non_registration_403_passthrough():
         )
         is None
     )
+
+
+def test_interactive_oauth_initial_failure_does_not_retry_loop(monkeypatch):
+    """Under force_interactive_oauth(), initial connection failure must fail fast
+    without burning reconnect retries that rotate PKCE state."""
+    from tools.mcp_tool import MCPServerTask
+    from tools.mcp_oauth import force_interactive_oauth
+
+    attempts = 0
+
+    async def _failing_http(self, config):
+        nonlocal attempts
+        attempts += 1
+        raise TimeoutError("HTTP read timeout")
+
+    monkeypatch.setattr(MCPServerTask, "_run_http", _failing_http)
+
+    task = MCPServerTask("oauth-srv")
+    task._auth_type = "oauth"
+
+    with force_interactive_oauth():
+        with pytest.raises(TimeoutError):
+            asyncio.run(task.start({"url": "https://mcp.example.com", "auth": "oauth"}))
+
+    assert attempts == 1
+
+
+def test_authorization_code_result_compatibility():
+    """_authorization_code_result must return a 2-tuple under mcp < 2.0 or AuthorizationCodeResult under mcp >= 2.0."""
+    from tools.mcp_oauth import _authorization_code_result
+
+    res = _authorization_code_result("test_code", "test_state", "test_iss")
+    if isinstance(res, tuple):
+        # Under mcp < 2.0, must unpack cleanly as 2-tuple
+        code, state = res
+        assert code == "test_code"
+        assert state == "test_state"
+    else:
+        assert res.code == "test_code"
+        assert res.state == "test_state"
+        assert res.iss == "test_iss"
+
+
+
--- a/tools/mcp_oauth.py
+++ b/tools/mcp_oauth.py
@@ -365,6 +365,12 @@ def force_interactive_oauth():
         _oauth_interactive_forced.reset(token)
 
 
+def _is_interactive_forced() -> bool:
+    """Return True if interactive OAuth was explicitly forced (e.g. mcp login / dashboard)."""
+    return bool(_oauth_interactive_forced.get())
+
+
+
 @contextmanager
 def suppress_interactive_oauth():
     """Disable stdin-based OAuth prompts for the current execution context.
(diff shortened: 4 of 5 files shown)


Enough1122:

**Overall:** Fixes two MCP OAuth reliability gaps: connect timeout not reaching transports (#98118) and retry loop rotating PKCE state during interactive login (#98118/#73997), plus SDK version compatibility.

**What it does**
- `hermes_cli/mcp_config.py:_probe_single_server` normalizes `config["connect_timeout"]` (was only set when `raw_timeout` existed; now also coerces explicit `connect_timeout` param, fallback 30s), then `config = dict(config); config["connect_timeout"]=connect_timeout` propagated to `_connect_server` so HTTP/SSE/stdio respect the 315s login timeout instead of 60s `streamable_http_client` default. Test `test_probe_propagates_connect_timeout_to_config` verifies 315.0 forwarded.
- `tools/mcp_oauth.py` adds `_is_interactive_forced()` (bool of `_oauth_interactive_forced`) and `_LegacyAuthCodeResult(tuple)` (compat `(code,state)` tuple with `.code/.state/.iss`) so `_authorization_code_result` returns tuple-like under `mcp<2.0` without losing `iss`.
- `tools/mcp_tool.py:MCPServerTask.run` adds single-flight gate: when `force_interactive_oauth()` is active (`hermes mcp login` / Dashboard), an initial failure (`not _ever_connected`) that is `TimeoutError`/etc. fails fast (`self._error=exc; self._ready.set(); return`) without retry loop that would re-enter `_run_http`, rotate PKCE `state`, print duplicate prompts, and collide ports. Test `test_interactive_oauth_initial_failure_does_not_retry_loop` asserts `attempts==1`.
- `_authorization_code_result` compat test ensures tuple-vs-object unpacking works.

**Non-blocking notes**
- `config = dict(config)` shallow copy — preserves caller's dict, good. If `config` contained nested mutable (headers), shallow copy is intentional (not deep) — those are not mutated here.
- `_is_interactive_forced` broad `except Exception: is_interactive_forced=False` — defensive; if contextvar lookup fails, falls back to normal retry behavior, safe.
- Gate only for `not _ever_connected` (initial connection) — once ever connected, resilience retries resume normally; correct scoping.
- `_LegacyAuthCodeResult` inherits tuple `(code,state)` so `code, state = result` works under both SDK versions; `iss` stored as `_iss` attr — check that SDK 2.0 path `AuthorizationCodeResult(code=..., state=..., iss=...)` is still used when available (it is, `try: from mcp.shared.auth import AuthorizationCodeResult` succeeds).

Well-targeted login robustness.

liyangbing:
<p>The change addresses the right failure boundary: <code>force_interactive_oauth()</code> now makes an initial connection failure fail fast, so <code>MCPServerTask.run()</code> does not re-enter the HTTP flow and mint a second PKCE state while the terminal is waiting for the first callback. Propagating the effective <code>connect_timeout</code> into the copied server config also keeps the transport timeout aligned with the browser-assisted flow.</p>

<h3>Suggested contract</h3>
<ul>
<li>Keep the interactive authorization attempt single-flight: one displayed authorization URL must correspond to one verifier/state pair until success, cancellation, or an explicit restart.</li>
<li>Preserve the fail-fast behavior only for an explicitly forced interactive flow; background/reconnect paths should retain their existing retry classification and must not inherit this gate accidentally.</li>
<li>Ensure the effective timeout is applied consistently to HTTP and SSE transports, while invalid or missing values continue to use the documented fallback and never produce an unbounded wait.</li>
<li>Clear or invalidate the pending callback result on success, mismatch, timeout, and cancellation so a late callback cannot complete a later login attempt.</li>
<li>Keep state/verifier material out of logs and error text; expose only the provider/server identity and a redacted failure category.</li>
</ul>

<h3>Regression matrix</h3>
<ul>
<li>forced interactive OAuth with an initial timeout, connection refusal, and authentication failure: exactly one attempt and the original error reaches the caller</li>
<li>interactive HTTP and SSE flows with an explicit timeout, including malformed and below-minimum values</li>
<li>callback success, state mismatch, callback replay, cancellation, and a second login after the first flow terminates</li>
<li>non-interactive startup and ordinary reconnects: existing retry/backoff behavior remains unchanged</li>
<li>an installed MCP SDK with <code>AuthorizationCodeResult</code> and an older SDK using the two-tuple compatibility path</li>
</ul>

<p>The two focused tests in this PR establish the most important regression points: timeout propagation and one-attempt failure under <code>force_interactive_oauth()</code>. The callback lifecycle and non-interactive retry boundaries are the remaining high-value cases to keep explicit.</p>

PR #67647, opened 2026-07-19
feat(credentials): configure dead manual pruning

DeadlySilent:
## What does this PR do?

Makes cleanup of manual credentials marked `dead` configurable instead of hard-coding removal after 24 hours.

The historical behavior remains the default: cleanup is enabled with a 24-hour TTL. Operators can disable cleanup to retain tombstones until explicit re-authentication/removal, or choose a bounded custom TTL. The CLI and dashboard now expose display-safe dead status details and guidance while keeping raw credentials out of API responses and logs.

Credential-pool dashboard reads and mutations also honor the selected management profile. Config writes use a cross-process lock around the partial read/merge/write transaction so unrelated concurrent configuration is preserved.

## Related Issue

Fixes #54487

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Add `credential_pool.prune_dead_manual_entries` and `credential_pool.dead_manual_prune_ttl_hours` with backward-compatible defaults (`true`, `24`).
- Validate TTLs strictly and bound them to 0–8,760 hours; disabling pruning provides indefinite retention.
- Keep dead credentials out of rotation while honoring the configured tombstone retention policy.
- Show canonical dead reason, representable UTC timestamp, and re-auth/remove guidance in `hermes auth list`.
- Add dashboard controls for cleanup policy and display-safe dead credential metadata.
- Scope credential-pool GET/PUT/POST/DELETE operations to the selected management profile.
- Serialize config writes across processes, persist only the partial `credential_pool` section, and return reloaded effective settings.
- Document the policy and add Python/API/frontend regression coverage for defaults, disabled cleanup, custom thresholds, invalid input, concurrent config preservation, profile isolation, timestamp validation, and API/log redaction.

## How to Test

1. Run `pytest tests/ -q`.
2. Run `npm --workspace web run typecheck`, `npm --workspace web run test`, `npm --workspace web run lint`, and `npm --workspace web run build`.
[...]

PR #98142, opened 2026-08-29
feat(credentials): configure dead manual pruning

DeadlySilent:
## Summary

- Restores the work from closed PR #67647 as a clean replacement against current `main`.
- Adds profile-scoped configuration for retention of dead manual credential tombstones.
- Preserves historical defaults: pruning enabled with a 24-hour TTL; invalid TTL values safely fall back to that policy.
- Keeps automatic pruning limited to manual credentials; seeded/singleton credentials remain protected.

## Verification

- `scripts/run_tests.sh tests/agent/test_credential_pool.py tests/hermes_cli/test_config.py -q`
  - 172 passed, 0 failed, 4 Windows-only skips
- `ruff check` passed for all changed Python and test files.
- `git diff --check origin/main...HEAD` passed.
- Independent read-only publication review: PASS.

## Attribution / prior PR

This replaces closed PR #67647 after its old PR record became detached during an authorship repair. The replacement contains only current-main-compatible, generic changes. All commits are authored and committed by `DeadlySilent`.

### DIFF
--- a/agent/credential_pool.py
+++ b/agent/credential_pool.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import math
 import os
 import random
 import threading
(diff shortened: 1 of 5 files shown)


Enough1122:

**Overall:** Makes dead-manual credential tombstone retention configurable per-profile, restoring revived work from #67647 with safe defaults (24h prune) and strict validation.

**What it does**
- `agent/credential_pool.py` adds `MAX_DEAD_MANUAL_PRUNE_TTL_HOURS = 24*365` and `get_dead_manual_prune_settings() -> (enabled,bool, ttl_seconds)` reading `credential_pool.prune_dead_manual_entries` (bool) and `dead_manual_prune_ttl_hours` (numeric, rejects bool, non-finite, out-of-range → fallback to 24h). Malformed config retains historical cleanup rather than disabling it.
- `_available_entries` lazily loads settings once (`dead_manual_prune_settings is None` → call helper) and gates pruning: `if prune_enabled and dead_at and now-dead_at > prune_ttl_seconds` (previously unconditional 24h). Seeded/singleton entries still never pruned (separate branch).
- `hermes_cli/config_defaults.py` adds `DEFAULT_CONFIG["credential_pool"] = {"prune_dead_manual_entries": True, "dead_manual_prune_ttl_hours": 24}`.
- Docs `website/docs/user-guide/features/credential-pools.md` and tests: profile-scoped policy (retain vs prune homes), non-finite fallback, default config exposure, existing 24h pruning still works.

**Non-blocking notes**
- `ttl_hours * 3600.0` could overflow for huge values but capped by `<= 8760` hours (31536000s) — safe.
- `isinstance(raw_ttl_hours, bool)` check needed because `bool` is subclass of `int` — correctly rejects `True` (1) as TTL. Good.
- `dead_manual_prune_settings` lazily cached per `_available_entries` call — if multiple dead entries exist, settings loaded once; correct.
- `get_dead_manual_prune_settings` calls `_load_config_safe()` each time `_available_entries` runs with a dead entry — could add `load_config()` overhead in hot path, but `_available_entries` is infrequent (pool selection), acceptable.

Clean configurable retention.

Issue #92554, opened 2026-08-22
Writing config.yaml destroys all user comments (config set strips them; plugins enable replaces them with default boilerplate)


hubbadubbadubdab:
### Summary

Any `hermes` command that writes `config.yaml` re-serializes it from the parsed dict, so **every user comment in the file is destroyed** — and Hermes' own default commentary block is written in its place. Comments in `config.yaml` are the natural place to record *why* a setting is what it is, and they do not survive routine CLI use.

### Reproduction

Fresh `HERMES_HOME` with a hand-written `config.yaml`:

```yaml
# TOP COMMENT — must survive
model:
  provider: test
plugins:
  # rationale for the enabled list
  enabled: []
hooks:
  pre_tool_call:
    # why this hook exists
    - command: /bin/echo
      timeout: 10
```

Then:

```
$ HERMES_HOME=$R hermes hooks doctor    # read-only control
   comments: 3  (unchanged)
$ HERMES_HOME=$R hermes plugins list    # read-only control
   comments: 3  (unchanged)
$ HERMES_HOME=$R hermes plugins enable demo
   comments: 36 — but all THREE originals are gone; the 36 are Hermes' default template comments
$ HERMES_HOME=$R hermes config set streaming.enabled true
   comments: 0
```

`hermes config set` strips everything. `hermes plugins enable` strips the user's comments and appends the stock "── Security ──" / "── Fallback Model ──" boilerplate, so a naive count of `#` lines goes *up* while the actual content is lost.

Read-only commands (`hooks doctor`, `plugins list`) correctly leave the file alone, so this is specific to the write path.

### Observed in practice

On a real install this removed 93 comments in a single session — including rationale recording *why* certain toolsets had been narrowed for security. The values were all preserved correctly; only the reasoning was lost, and silently, with no warning that the file had been rewritten.

### Expected

Preserve comments across writes — e.g. `ruamel.yaml` round-trip mode, which is the usual answer for a config file humans are expected to hand-edit. Failing that, warn on the first destructive write, or write changes surgically rather than re-serializing the whole document.

### Why it matters

The docs encourage hand-editing `config.yaml`, and several settings genuinely need a comment to be intelligible later (why a `platform_toolsets` entry was trimmed, why a hook is `fail_closed`, why a provider was pinned). Losing that on the next `hermes plugins enable` is a quiet documentation-integrity bug: nothing errors, nothing is flagged, and the loss is only noticed much later when someone asks why a setting is the way it is.

[...]

Issue #63039, opened 2026-07-12
config: set_config_value drops comments/formatting — diverges from save_config_value


crazythinking:
## Problem

`hermes config set` silently strips all comments, blank lines, and formatting from `config.yaml` on every write.

## Root Cause

Two separate implementations exist for writing config, with different behavior:

| Function | Module | Serializer | Preserves comments? |
|---|---|---|---|
| `save_config_value` | `cli.py` | `ruamel.yaml` (round-trip) | ✅ |
| `set_config_value` | `hermes_cli/config.py` | `PyYAML` (`yaml.dump`) | ❌ |

The `hermes config set` CLI command dispatches to `set_config_value` → `atomic_yaml_write`, which uses PyYAML and drops all comments and formatting.

`cli.py:save_config_value` (used by `/model`, `/approvals`, `/skin`, etc.) was updated in commit `85383c636` to use `atomic_roundtrip_yaml_update` with `ruamel.yaml` — but the older `set_config_value` was never updated.

Additionally, `hermes update` triggers `migrate_config` → `save_config` which rewrites the entire config in `DEFAULT_CONFIG` key order (defined at `config.py:976`), causing sections like `fallback_model` and `custom_providers` (not in DEFAULT_CONFIG) to be shuffled to the end of the file.

## Steps to Reproduce

1. Add comments to `~/.hermes/config.yaml`:
```yaml
# My custom auxiliary setup
auxiliary:
  model: gpt-4o-mini
```
2. Run `hermes config set auxiliary.model gpt-4o`
3. Inspect `config.yaml` — the comment is gone

## Proposed Fix

Replace `fast_safe_load` + `_set_nested` + `atomic_yaml_write` in `hermes_cli/config.py:set_config_value` with `ruamel.yaml` round-trip read + write, matching the approach already used by `cli.py:save_config_value`.

## Related

The `atomic_roundtrip_yaml_update` helper in `utils.py` could be extended to support numeric list indices (currently only dict nesting), which would allow `set_config_value` to use it directly instead of inlining ruamel.yaml boilerplate.

sidharthgehlot:
I reproduced the divergence between `set_config_value` and the existing round-trip YAML path and am working on a focused test-first fix. I found no open competing PR immediately before starting.

teknium1:
Fixed on main by #118918 — merge commit `43504da0e5993750e1e5816f47365c43c29f79af`.

[...]

Issue #50698, opened 2026-06-22
hermes config set strips all comments and commented-out template blocks from config.yaml


notwitcheer:
## summary

`hermes config set <key> <value>` serializes `config.yaml` back out through a comment-eating YAML writer. it updates the targeted key correctly and keeps the active settings, but every comment and every commented-out template section is deleted on save. anyone who keeps the shipped documented scaffolding (for example the commented `fallback_model:` example block) or their own inline notes loses it silently, which is why the community has written a PSA recommending people hand-edit the config files instead.

## steps to reproduce (v0.17.0)

1. start from a `config.yaml` that contains comments and commented-out template sections (the default shipped config does, e.g. the trailing "Fallback Model" block).
2. run any single edit, e.g. `hermes config set memory.memory_char_limit 2300`.
3. inspect `config.yaml`: the targeted key is updated, but all comments and the commented-out blocks are gone.

## observed

clean repro on a copied config in a temporary `HERMES_HOME`:

- before: 677 lines, 71 top-level keys.
- after `config set memory.memory_char_limit 2300`: 655 lines, 71 top-level keys.
- the field updated correctly (`2200` to `2300`) and no active setting was lost, but the entire trailing 22-line commented "Fallback Model" block was deleted: the documentation comments plus the ready-to-uncomment `fallback_model:` example (the provider list and the sample config).

## expected

`config set` preserves comments and commented-out config on save, by round-tripping through a comment-preserving YAML library such as ruamel.yaml.

## impact

- silent loss of inline docs and commented scaffolding, including the shipped `fallback_model:` example users are meant to uncomment to enable provider failover.
- data-loss-adjacent, and a docs/trust gap: the community is now routing around the CLI by hand-editing.

## environment

- Hermes Agent v0.17.0 (2026.6.19), macOS. reproduced against a copied config in a temporary `HERMES_HOME` (non-destructive).

## reference

community PSA, r/hermesagent, 2026-06-22: https://www.reddit.com/r/hermesagent/comments/1ubyu8w/psa_how_to_safely_edit_hermes_config_files/

alt-glitch:
[...]

PR #98143, opened 2026-08-29
fix(config): stop `config set` from deleting config.yaml comment blocks

MKappes:
## Summary

`hermes config set <any key>` permanently deletes the commented-out
documentation blocks (**Security**, **Fallback Model**) from the bottom of the
user's `config.yaml`.

There are two write paths and only one of them is correct:

| Writer | `atomic_yaml_write` call | Comments |
|---|---|---|
| `save_config` (config.py:4109) | passes `extra_content=` | preserved |
| `set_config_value` (config.py:5788) | **no `extra_content`** | **deleted** |

`atomic_yaml_write` rewrites the whole document. The blocks are comments, not
mapping entries, so they are not part of the round-tripped data and nothing
downstream can restore them. The loss is permanent and silent.

### Reproduction

```bash
grep -c "── Security ──\|── Fallback Model ──" ~/.hermes/config.yaml   # 2
hermes config set agent.max_turns 200
grep -c "── Security ──\|── Fallback Model ──" ~/.hermes/config.yaml   # 0
```

Found in the wild via `hermes config set mcp_servers.gitlab.enabled false`,
which removed 38 lines of documentation the user had never touched.

## Fix

Extract the block-selection logic into `_optional_comment_sections()` and call
it from **both** writers. One source of truth rather than a duplicated block
that a second caller can silently forget.

No behavior change to `save_config` — the helper is a literal extraction of the
logic that was already there.

## Test plan

Four tests in `tests/hermes_cli/test_config_set_preserves_comments.py`,
written as behavior contracts rather than snapshots:

- comment blocks survive a `config set`
- repeated sets never duplicate them
- a genuinely configured `fallback_model` still suppresses its own placeholder
- unit-level invariants for the helper (mapping form, list form, incomplete
  entry, explicit `security.redact_secrets`)

Verified RED/GREEN: **3 of the 4 fail against the unpatched writer** and pass
with it. The fourth is a pure-unit helper contract and is correctly
insensitive to the writer.

Regression run against a clean `origin/main` worktree for comparison:

```
tests/hermes_cli/ -k "mcp or config"
  baseline (clean HEAD):  35 failed, 862 passed
  with this change:       35 failed, 870 passed
```

Identical failures, same test names — all pre-existing cross-test pollution
(they pass individually). Zero regressions, +8 passing.

Also verified end-to-end against a real install: after the fix,
`hermes config set agent.max_turns 200` produces a one-line diff (the value)
with both comment blocks intact.

### DIFF
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -4013,6 +4013,36 @@ _COMMENTED_SECTIONS = """
 """
 
 
+def _optional_comment_sections(normalized: Any) -> list:
+    """Return the commented-out doc blocks that belong at the end of config.yaml.
+
+    These blocks document features that are off by default (secret redaction
+    tuning, ``fallback_model``) and are appended as ``extra_content`` by the
+    writer. They are emitted only while the corresponding section is unset, so
+    a user who configures ``fallback_model`` stops seeing the placeholder.
+
+    Shared by :func:`save_config` and :func:`set_config_value`. Before this was
+    factored out, ``set_config_value`` called ``atomic_yaml_write`` with no
+    ``extra_content``, so a single ``hermes config set`` silently deleted these
+    comment blocks from the bottom of the user's config.yaml.
+    """
+    parts: list = []
+    sec = normalized.get("security", {})
+    if not sec or (isinstance(sec, dict) and sec.get("redact_secrets") is None):
+        parts.append(_SECURITY_COMMENT)
+    fb = normalized.get("fallback_model", {})
+    fb_is_valid = False
+    if isinstance(fb, list):
+        fb_is_valid = any(
+            isinstance(e, dict) and e.get("provider") and e.get("model") for e in fb
+        )
+    elif isinstance(fb, dict):
+        fb_is_valid = bool(fb.get("provider") and fb.get("model"))
+    if not fb_is_valid:
+        parts.append(_FALLBACK_COMMENT)
+    return parts
+
+
 def save_config(
     config: Dict[str, Any],
     *,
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Fixes `hermes config set` silently deleting trailing commented doc blocks (Security + Fallback Model) from `config.yaml`. Single shared helper restores `save_config` parity.

**What it does**
- `hermes_cli/config.py` extracts `_optional_comment_sections(normalized)` — returns `[ _SECURITY_COMMENT ]` if `security.redact_secrets is None` and `[ _FALLBACK_COMMENT ]` if `fallback_model` incomplete (dict missing provider/model or list with no valid entry). Factoring out previously duplicated logic from `save_config`.
- `save_config` now `parts = _optional_comment_sections(normalized)` (behavior-preserving refactor).
- `set_config_value` now rebuilds `_extra = _optional_comment_sections(user_config)` and calls `atomic_yaml_write(..., extra_content="".join(_extra) if _extra else None, sort_keys=False)` — previously `atomic_yaml_write(config_path, user_config, sort_keys=False)` with no `extra_content`, so `atomic_yaml_write`'s full-document rewrite dropped comments permanently (they are not in parsed mapping). Mirrors `save_config` writer path.
- Tests `test_config_set_preserves_comments.py`: single set preserves both blocks, 3× set not duplicate (count==1), configured fallback suppresses its placeholder, unit contract for helper.

**Non-blocking notes**
- `_optional_comment_sections` checks `isinstance(sec, dict)` but earlier `if not sec` already handled empty/non-dict; fallback to `sec.get` after `not isinstance` guard would be safer, but `sec` is normally dict from defaults/merge — low risk.
- `extra_content="".join(_extra) if _extra else None` — passing `None` when no blocks omits append; `_extra` empty when both configured →no trailing comments, correct.
- `set_config_value` reads `user_config` (raw without defaults merged?) then checks `fallback_model` validity there — if user has no fallback but defaults provide one via merge elsewhere, placeholder still shown in file — intended (placeholder is doc for unset feature).

Well-factored bug fix.

teknium1:
Thank you for working on this — the class this PR fixes landed on main in #118918 (`43504da0e5993750e1e5816f47365c43c29f79af`), which routes every `config.yaml` writer (not only the one this PR targets) through a single ruamel round-trip writer, `hermes_cli.config.atomic_config_write`, and adds a regression test plus a CI lint so no writer can pick PyYAML again. You are credited in the PR body and on #92554 / #63039 / #50698. Closing as superseded; if a case you covered is not preserved on current main, please reopen with the config snippet.

PR #98144, opened 2026-08-29
fix(mcp): don't report a post-auth session failure as "Authentication failed"

MKappes:
## Summary

`_reauth_oauth_server` wraps the whole probe in a single `except` that prefixes
**every** exception with `Authentication failed:` (`mcp_config.py:897`). That is
wrong whenever the OAuth round trip actually succeeded and the failure came
later, in the MCP session.

The result is a message that names the one stage that *did* work, so the user
re-runs the browser flow against a server that has already issued them a valid
token.

### Real-world reproduction (GitLab)

`hermes mcp login gitlab` completed the browser flow and cached a valid token:

```json
// ~/.hermes/mcp-tokens/gitlab.json
{ "access_token": "<64 chars>", "scope": "mcp", "refresh_token": "<64 chars>" }
```

Then `POST /api/v4/mcp` with that token returned:

```
HTTP 404  {"message":"404 Not Found"}
ratelimit-name: throttle_authenticated_api      <-- token WAS accepted
```

GitLab gates its MCP endpoint behind GitLab Duo settings that only exist on
**group** namespaces, so a personal namespace gets a 404 after a fully
successful authorization. The MCP SDK surfaces this as the generic
`Session terminated` (`streamable_http.py:519`), and Hermes printed:

```
x Authentication failed: Session terminated
```

Two full browser round trips were spent before the token on disk was noticed.

## Fix

The discriminator already existed three lines above — `_oauth_tokens_present`,
added for exactly this class of confusion. Consult it in the error path.

**Before**

```
x Authentication failed: Session terminated
```

**After**

```
x Connected, but the MCP session failed: Session terminated

  Authorization SUCCEEDED — a token is cached at ~/.hermes/mcp-tokens/gitlab.json.
  Re-running `hermes mcp login gitlab` will not help; the failure is after auth.
  Common causes: the server requires a plan/feature/permission your account
  lacks, the endpoint is disabled for your org/namespace, or the URL is wrong.
  Check the vendor's MCP docs for prerequisites, then `hermes mcp test gitlab`.
```

A genuine auth failure (no cached token) keeps the original message unchanged.

`_token_path_hint` is display-only and falls back to the conventional path; it
never influences a control decision.

## Test plan

Four tests in `tests/hermes_cli/test_mcp_login_post_auth_failure.py`:

- token present + probe raises → **not** labelled an auth failure, real error
  still shown, guidance says re-running login won't help
- no token + probe raises → original "Authentication failed" preserved, and no
  misleading "auth succeeded" guidance
- `_token_path_hint` degrades to a string, never raises
- the success path is unchanged

Verified RED/GREEN: the first test fails against the unpatched code with
exactly the string from the field report:

```
AssertionError: a post-auth session failure must not be labelled an auth failure
assert 'Authentication failed' not in 'Authentication failed: Session terminated'
```

Regression run against a clean `origin/main` worktree:

```
tests/hermes_cli/ -k "mcp or config"
  baseline (clean HEAD):  35 failed, 862 passed
  with this change:       35 failed, 870 passed
```

Identical failures, same test names — all pre-existing cross-test pollution
(they pass individually). Zero regressions.

### DIFF
--- a/hermes_cli/mcp_config.py
+++ b/hermes_cli/mcp_config.py
@@ -417,6 +417,24 @@ def _oauth_tokens_present(name: str) -> bool:
         return True
 
 
+def _token_path_hint(name: str) -> str:
+    """Best-effort human-readable path to the cached OAuth token for ``name``.
+
+    Used in error messages so the user can see for themselves that auth
+    succeeded. Falls back to the conventional location if the storage class
+    can't be interrogated — this is display text, never a control decision.
+    """
+    try:
+        from tools.mcp_oauth import HermesTokenStorage
+
+        path = getattr(HermesTokenStorage(name), "token_path", None)
+        if path:
+            return str(path)
+    except Exception as exc:  # pragma: no cover — defensive
+        logger.debug("Could not resolve token path for '%s': %s", name, exc)
+    return f"~/.hermes/mcp-tokens/{name}.json"
+
+
 def _unwrap_exception_group(exc: BaseException) -> Exception:
     """Extract the root-cause exception from anyio TaskGroup wrappers.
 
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Prevents misleading "Authentication failed" message when OAuth succeeded but subsequent MCP session setup failed. Observed with GitLab 404 on `POST /api/v4/mcp` after valid `scope: mcp` token was cached, causing users to retry browser flow against an already-issuing server.

**What it does**
- `hermes_cli/mcp_config.py` adds `_token_path_hint(name)` (resolves `HermesTokenStorage(name).token_path` or falls back to `~/.hermes/mcp-tokens/{name}.json`, debug-log on exception).
- `hermes_cli/mcp_config.py:_reauth_oauth_server` after `except Exception:` (whole probe): if `_oauth_tokens_present(name)` (token on disk), reports `Connected, but the MCP session failed: {humanized or exc}` + info `Authorization SUCCEEDED — a token is cached at {path}... will not help` + common causes (plan/feature, endpoint disabled, wrong URL) + `hermes mcp test <name>` guidance; returns `False`. Else (no token) retains original `Authentication failed: ...` path.
- Tests `test_mcp_login_post_auth_failure.py`: post-auth shows session failure not auth, genuine shows auth, `token_path_hint` never raises, success path unchanged.

**Non-blocking notes**
- Discriminator `_oauth_tokens_present` checks file existence — if token was just written but not yet flushed to disk (concurrent), could briefly misclassify post-auth as genuine auth failure, but window is tiny and retry will then show SUCCEEDED — acceptable.
- `_token_path_hint` instantiates `HermesTokenStorage(name)` just for `token_path` — ensure no side effects (file creation) in that constructor; assuming path property is pure.
- Message `Connected, but the MCP session failed` correctly shifts attribution away from OAuth; ensure UI (color) uses `_error` level still visible.

Helpful UX fix.

Issue #93258, opened 2026-08-23
[Bug]: Windows: hermes update exits 1 — fleet check flags STALE pid of an already-exited gateway


jrleal10:
## Bug Description

`hermes update --yes` on Windows with **two live profile gateways** completed the code pull (333 commits, `b2c4f1f376` → `739bc555b1`) but exited **1**. The only failure is the post-update fleet check, which reported:

```
Fleet version check:
  ✗ joao_pessoal (pid 40652) @ b2c4f1f3 — STALE (pre-update code)
```

PID **40652 was already gone**. `gateway.lifecycle.json` for that profile recorded:

```json
{"phase": "exited", "pid": 40652, "exit_code": 0, "exit_reason": "graceful_shutdown",
 "exited_at": "2026-08-23T22:12:15.530209+00:00"}
```

The updater itself had paused both profiles at the start (`✓ Paused gateway profile(s): arthur_tutor, joao_pessoal`). After "Update complete" it printed `✓ Restarting Windows gateway profile(s): arthur_tutor, joao_pessoal`, then immediately ran the fleet check against the **old** PID instead of the process it had just spawned (or instead of noticing 40652 was dead).

~50s later `joao_pessoal` was up on **new** PIDs (23204/27284, `--replace`) and Telegram connected. The exit-1 is a **false stale**.

Related: #91675 (multi-profile Windows update / cold-start). This is a new failure mode on a run where we did **not** stop the gateways ourselves — we left that to the updater, as requested.

## Steps to Reproduce

1. Windows 11, two profiles with live gateways (`arthur_tutor` Discord, `joao_pessoal` Telegram).
2. Leave both running. Do **not** `gateway stop`. A 5-minute local watchdog may exist for one profile (see below).
3. `python -m hermes_cli.main update --yes` (same updater as `hermes update`; we avoid `hermes.exe` because of the known Windows self-lock).
4. Observe pause → pull → "Restarting … arthur_tutor, joao_pessoal" → fleet check `✗ joao_pessoal (pid <old>) STALE` → process exit 1.
5. `Get-Process -Id <old>` → not found. New gateway PIDs appear shortly after; Telegram/Discord connect.

## Expected Behavior

- Fleet check should not flag a PID that is no longer running.
- After the updater claims it restarted a profile, the check should look at the **new** PID (or wait for it), not the pre-pause one.
- Exit 0 when code is updated and the only "stale" PID is already dead / being replaced.
[...]

PR #98145, opened 2026-08-29
fix(update): reject recycled gateway state PIDs

wang-yzh:
## Summary

- prevent `hermes update` from counting a stale `gateway_state.json` PID as a live gateway after the PID has been reused by an unrelated process
- apply the same runtime identity validation to both the pre-update inventory and post-update fleet verification paths
- add coverage for unrelated-process PID reuse and cross-profile gateway PID reuse

## Root cause

The `gateway_state.json` fallback in `collect_runtime_inventory()` only checked whether the recorded PID existed. `collect_fleet_versions()` used `runtime_status_pid_is_live()`, which verifies PID/start-time identity but not that a readable live command line is actually a Hermes gateway for the expected profile.

A long-dead profile record on macOS contained PID 1231. That PID had since been reused by `SafariWidgetExtension`, so `hermes update` printed:

```
Fleet: 1 running service(s) across profiles: work
```

No Hermes gateway was running.

## Fix

Reuse the existing `get_runtime_status_running_pid(record, expected_home=home)` validator in both update paths. It checks:

- the recorded PID exists and is not a zombie
- the recorded process start time still matches when available
- a readable live command line is a Hermes gateway runtime
- the gateway command belongs to the expected profile

The control-socket path remains authoritative and unchanged. Platforms where the command line cannot be read retain the validator's existing conservative metadata fallback.

## Test plan

```
venv/bin/python3 -m pytest -o 'addopts=' -q \
  tests/hermes_cli/test_update_inventory.py \
  tests/hermes_cli/test_update_receipt.py \
  tests/hermes_cli/test_fleet_matrix_down_state.py \
  tests/hermes_cli/test_gateway_runtime_health.py
# 48 passed

venv/bin/python3 -m ruff check \
  hermes_cli/update_inventory.py \
  hermes_cli/update_receipt.py \
  tests/hermes_cli/test_fleet_matrix_down_state.py \
  tests/hermes_cli/test_update_inventory.py \
  tests/hermes_cli/test_update_receipt.py
# All checks passed

git diff --check
```

Live verification against the original stale state file now returns:

```
inventory_runtimes= []
fleet_versions= []
```

The broad `tests/hermes_cli/test_update*.py` run has pre-existing failures on this macOS checkout: an untouched `origin/main` worktree produced 627 passed, 3 skipped, 19 failed; this branch produced 628 passed, 3 skipped, 20 failed, with the one additional loop-heartbeat timing failure passing when rerun alone.

## Disclosure

This patch was prepared with Hermes Agent assistance and independently reviewed before submission.

### DIFF
--- a/hermes_cli/update_inventory.py
+++ b/hermes_cli/update_inventory.py
@@ -247,7 +247,7 @@ def collect_runtime_inventory() -> UpdatePlan:
     # --- per-profile gateways (PID files + runtime status stamps) ----------
     seen_pids: set[int] = set()
     try:
-        from gateway.status import _pid_exists, read_runtime_status
+        from gateway.status import get_runtime_status_running_pid, read_runtime_status
 
         for profile, home in profile_homes:
             # Prefer the gateway-owned control socket (#92091): identity
@@ -297,13 +297,10 @@ def collect_runtime_inventory() -> UpdatePlan:
             pid: Optional[int] = None
             code_sha = code_version = None
             if record:
-                try:
-                    pid = int(record.get("pid"))
-                except (TypeError, ValueError):
-                    pid = None
+                pid = get_runtime_status_running_pid(record, expected_home=home)
                 code_sha = record.get("code_sha")
                 code_version = record.get("code_version")
-            if pid is None or not _pid_exists(pid):
+            if pid is None:
                 continue
             seen_pids.add(pid)
             supervisor = _detect_supervisor_for_pid(
--- a/hermes_cli/update_receipt.py
+++ b/hermes_cli/update_receipt.py
@@ -343,7 +343,7 @@ def collect_fleet_versions(
         expected_sha = None
 
     try:
-        from gateway.status import read_runtime_status, runtime_status_pid_is_live
+        from gateway.status import get_runtime_status_running_pid, read_runtime_status
         from hermes_cli.profiles import (
             _get_default_hermes_home,
             _get_profiles_root,
@@ -405,7 +405,8 @@ def collect_fleet_versions(
                 pid = int(pid)
             except (TypeError, ValueError):
                 continue
-            if not runtime_status_pid_is_live(record):
+            live_pid = get_runtime_status_running_pid(record, expected_home=home)
+            if live_pid is None:
                 # Dead PID (or a live PID recycled by an unrelated process
                 # during the update's own churn — #93258): a DOWN row only
                 # when this exact pid was alive at update start AND the
@@ -437,6 +438,7 @@ def collect_fleet_versions(
                         }
                     )
                 continue
+            pid = live_pid
             code_sha = record.get("code_sha")
             if not code_sha or not expected_sha:
                 state = "unknown"
--- a/tests/hermes_cli/test_fleet_matrix_down_state.py
+++ b/tests/hermes_cli/test_fleet_matrix_down_state.py
@@ -27,6 +27,9 @@ def _setup(monkeypatch, tmp_path, record: dict):
         "hermes_cli.profiles._get_profiles_root", lambda: tmp_path / "no-profiles"
     )
     monkeypatch.setattr("gateway.control_socket.identify_gateway", lambda h, **k: None)
+    monkeypatch.setattr(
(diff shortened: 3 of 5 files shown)


Enough1122:

**Overall:** Hardens `hermes update` fleet inventory against PID reuse — stale `gateway_state.json` PID that now belongs to Safari/an unrelated gateway no longer counted as live.

**What it does**
- Replaces naive `_pid_exists(pid)` / `runtime_status_pid_is_live(record)` checks with `get_runtime_status_running_pid(record, expected_home=home)` which validates runtime identity: `kind`, `argv`, `gateway_state=="running"`, `start_time`, and cmdline `hermes gateway run` with correct `--profile` for named homes (via `_read_process_cmdline`). Returns `None` for recycled PID, else live PID.
- Applied in both `collect_runtime_inventory()` (pre-update inventory, `pid = get_runtime_status_running_pid(...); if pid is None: continue`) and `collect_fleet_versions()` (post-update verification, `live_pid = get...; if None` → dead/reused handling; `pid = live_pid`).
- Tests add `Safari` and `other-profile` reuse cases for both paths, mock `_read_process_cmdline` to return foreign cmdline.

**Non-blocking notes**
- Identity via cmdline is best-effort — a malicious process could spoof `hermes gateway run` argv, but threat here is accidental PID reuse, not adversarial spoofing; sufficient.
- `_read_process_cmdline` may fail (permission, `/proc` unavailable on macOS sandbox) — ensure helper returns empty/fallback and `get_runtime_status_running_pid` treats that as not-live (safe side: excludes runtime rather than including stale).
- Both `collect_runtime_inventory` and `collect_fleet_versions` now import `get_runtime_status_running_pid` — previous `_pid_exists` still mocked in old tests but no longer exercised; verify no dead import.

Correct hardening for PID-recycle during update churn (#93258).

sulee2019:
Independent confirmation on macOS — same failure shape as the Safari case in the description, different victim process.

**Environment**: Hermes v0.21.1, git install at `~/.hermes/hermes-agent` (HEAD `ee35a4624f`), desktop app with the exit-code hand-off.

**What happened**
- `~/.hermes/profiles/history-expert/gateway_state.json` still declared `"gateway_state": "running"` with `"pid": 26837`, last written **2026-06-25** — that gateway had died ungracefully ~2.5 months earlier.
- PID 26837 had since been reused by an unrelated process: `/System/Applications/Books.app/Contents/PlugIns/BooksThumbnail.appex/Contents/MacOS/BooksThumbnail` (Apple Books QuickLook extension).
- `hermes update` therefore planned it: `plan.runtimes` contained `{kind: gateway, profile: history-expert, pid: 26837, supervisor: manual}`. The restart phase cannot satisfy a `manual` runtime, so the receipt was written as `outcome: partial`, the updater exited 1, and the desktop app raised "Hermes update did not finish".
- It also left `~/.hermes/fleet_restart_pending` behind, which then pinned the "A previous `hermes update` pulled new code but did not restart running gateways" warning on every later CLI start.
- The update itself had actually succeeded: both real gateways reported the post-update SHA, and the only unhealthy thing in the fleet was the phantom.

**Where it comes from** (v0.21.1, and still on `main` at `073c57872`)
- `hermes_cli/update_inventory.py::_collect_gateway_runtimes` — the `gateway_state.json` fallback path does `if not _pid_exists(pid): continue`, so a recycled PID passes.
- `gateway.status.get_runtime_status_running_pid(record, expected_home=home)` rejects the same record via the start-time guard, which is what this PR switches to. The control-socket path is unaffected.

**Read-only reproduction**
`hermes update --plan` lists the phantom runtime under "Running services to restart" — no code changes or cleanup required to observe it.

**Workaround applied on the affected host**
Moved the stale `gateway_state.json` aside and cleared `fleet_restart_pending`; `hermes update --plan` then returns to a 4-item plan with no phantom, and the exit code returns to 0. That is per-machine cleanup only — the class of bug (any profile whose gateway dies ungracefully and whose recorded PID is later reused by another process) remains until this identity validation is applied to the inventory path.

Happy to test a patch on this host (Apple Silicon, launchd-supervised gateways + desktop hand-off) if that helps the review.

PR #98148, opened 2026-08-29
fix(context): classify OmniRoute combo limit overflow

anhtahaylove:
## What does this PR do?

Classifies OmniRoute''s combo context-window admission message as `context_overflow` even when a custom-provider wrapper drops the structured HTTP status code.

Without this pattern, Hermes classifies the wrapped exception as `unknown` / non-compressing and can abort a long-running session instead of entering the existing compact-and-retry recovery path.

Observed provider message:

```text
HTTP 400: Request requires approximately 278690 tokens, but the largest known context limit in this combo is 272000 tokens. Reduce or compact the request context.
```

## Related Issue

No existing matching issue or PR found.

## Type of Change

## Changes Made

- Add `"largest known context limit"` to `_CONTEXT_OVERFLOW_PATTERNS` in `agent/error_classifier.py`.
- Add a regression test for the message-only, no-`status_code` wrapper path in `tests/agent/test_error_classifier.py`.

## How to Test

1. Run `python -m pytest tests/agent/test_error_classifier.py -q`.
2. Run `python -m pytest tests/run_agent/test_overflow_overhead_aware_tokens.py -q`.
3. Verify a wrapped OmniRoute error classifies as `FailoverReason.context_overflow` with `retryable=True` and `should_compress=True`.

Observed locally after the fix: Hermes compacted an affected session from approximately 212k estimated tokens to 53.2k and the next provider request succeeded.

### Code

- [x] I''ve read the Contributing Guide
- [x] I searched for existing PRs to make sure this isn''t a duplicate
- [ ] I''ve run `pytest tests/ -q` and all tests pass (targeted suites run; full suite not run)
- [x] I''ve added tests for my changes
- [x] I''ve tested on my platform: Windows 11

- [x] Documentation update — N/A
- [x] `cli-config.yaml.example` update — N/A
- [x] `CONTRIBUTING.md` / `AGENTS.md` update — N/A
- [x] Cross-platform impact considered: provider-message classification is platform-independent
- [x] Tool descriptions/schemas update — N/A

## Screenshots / Logs

Targeted verification:

```text
tests/agent/test_error_classifier.py: 123 passed
tests/run_agent/test_overflow_overhead_aware_tokens.py: 7 passed
```


Enough1122:

**Overall:** One-line classifier fix for OmniRoute combo admission message that lacks a usable HTTP status after custom-provider wrapping, allowing compact-and-retry instead of aborting session.

**What it does**
- `agent/error_classifier.py` adds `"largest known context limit"` to the `context_overflow` message patterns (alongside `"maximum allowed input length"`). Observed wrapped message: `HTTP 400: Request requires approximately 278690 tokens, but the largest known context limit in this combo is 272000 tokens. Reduce or compact...` — previously `unknown`/non-compressing after wrapper dropped structured status.
- Test `test_omniroute_wrapped_context_limit_without_status_triggers_compression` asserts `classify_api_error` with that message (provider custom, no status_code) → `FailoverReason.context_overflow`, `retryable=True`, `should_compress=True`.

**Non-blocking notes**
- Pattern `"largest known context limit"` is sufficiently specific to OmniRoute combo message; low false-positive risk. Check that normal non-overflow messages don't contain it — unlikely.
- Classifier message is lowercased before check (`error_msg.lower()`), but pattern itself is lowercase — match is case-insensitive via lowercasing, correct.

Minimal, correct.

anhtahaylove:
Thanks — confirmed. The phrase is specific to OmniRoute combo admission overflow, classification already lowercases the message, and the regression pins `context_overflow` plus compact-and-retry semantics. I am keeping the change scoped as-is.

anhtahaylove:
Gentle nudge on this one — it has been open ~10 days with an AI review and my reply, but no maintainer decision yet. Sharing production evidence in case it helps triage, since the failure mode is worse than the one-line diff suggests.

**Before the patch** (`2026-08-29T14:12:58Z`, unpatched `error_classifier.py`) the wrapped OmniRoute message is not recognised as context overflow, so the run is abandoned outright:

```
📝 Error: HTTP 400: Request requires approximately 273875 tokens, but the largest
   known context limit in this combo is 272000 tokens. Reduce or compact the request context.
📋 Details: {'type': 'invalid_request_error', 'code': 'context_length_exceeded'}
❌ Non-retryable error (HTTP 400)
❌ Non-retryable client error (HTTP 400). Aborting.
   💡 This type of error won't be fixed by retrying.
⚠️  Skipping session persistence for large failed session to prevent growth loop.
```

Note the last line: the session is not just aborted, it is **dropped without persistence**, so the user loses the conversation. The overflow is only 1,875 tokens over a 272,000 limit — entirely recoverable by the compaction path that already exists.

**After the patch** (`2026-08-30T13:50:31Z`, same machine, same provider) the identical error shape is classified correctly and recovers:

```
⚠️  Context length exceeded, but provider did not report a max context length;
    keeping context_length at 272,000 tokens and compressing.
🗜️ Context too large (~197,208 tokens) — compressing (2/3)...
🗜️ Compressed ~209,836 → ~128,861 tokens, retrying...
```

The `code: 'context_length_exceeded'` field is present in `Details` both times — the provider is telling us exactly what is wrong; only the message-pattern branch misses it, because the custom-provider wrapper drops the structured status before `classify_api_error` sees it.

For scale: 31 occurrences in ~10 days of a single desktop log, across two separate days of normal use.

Happy to add anything that would help this land — a second regression test for the `code`-based path, or narrowing the pattern further if the concern is over-matching. If the phrase itself is considered too provider-specific to hard-code, I am equally happy to rework this to key off `details.code == 'context_length_exceeded'` instead, which would be provider-agnostic and cover the same case. Just let me know which direction you prefer.

anhtahaylove:
Independent corroboration of the root cause here, which I think is the strongest argument for this PR: #105463 (opened 2026-09-08, unrelated author, different symptom) arrived at the same ordering defect from the opposite direction. Their inline comment:

> `_by_status` runs before `_by_error_code`, so the 400 [...]

They hit it classifying local-inference memory-ceiling rejections; this PR hits it classifying OmniRoute context overflow. Two contributors independently finding that `_classify_400` decides from the message before the structured `error.code` is ever consulted is a signal the ordering itself is the bug, not either symptom.

The two changes are complementary, not competing — I checked the diffs line by line:

- #105463 edits `_MESSAGE_HEAD_RULES` / adds `_MEMORY_CEILING_PATTERNS`
- this PR edits the `_ERROR_CODE_VERDICTS` mapping and the structured-code branch in `_classify_400`

No shared lines, so they merge in either order. Worth noting their fix relies on the same precedence being correct, so landing this one strengthens theirs.

Status recap, since this has been open ~10 days: one AI review, no human review yet. Local suite is `137/137` on the classifier tests, including three added regression tests (structured `context_length_exceeded`, a malformed-body guard, and the OmniRoute message-only fallback for wrappers that drop the structured body). The RED/GREEN was verified against a clean `origin/main` copy of `error_classifier.py`, not just against my working tree — my first attempt at the RED run passed spuriously because the reset had not actually taken, so I re-did it with `git show origin/main:agent/error_classifier.py`.

The production evidence remains as posted earlier: 31 occurrences of the OmniRoute overflow message in `desktop.log`, each currently classified non-retryable and ending the turn instead of triggering compaction.

anhtahaylove:
Status update rather than a nudge: this is now unblocked on my side.

- **Rebased onto current main** (`875a096939`). It had gone CONFLICTING against b7e4712029 (#105463, the memory-ceiling classifier that landed 09-09) — that change touches the same `_ERROR_CODE_VERDICTS` entry and the same insertion point in `_classify_400`. `mergeable` is back to MERGEABLE.
- **The two changes are complementary, not competing.** `_MEMORY_CEILING_ERROR_CODES` (`prefill_memory_exceeded`, …) and `_CONTEXT_OVERFLOW_ERROR_CODES` (`context_length_exceeded`, `max_tokens_exceeded`) are disjoint sets, so both code checks now sit side by side and order between them does not affect behaviour. Worth noting #105463 reached the same `_by_status`-before-`_by_error_code` ordering defect from an unrelated symptom — this PR is the other half of it.
- **Full classifier suite passes: 156/156**, covering both #105463's tests and the two regressions added here.
- **Closed #107898** (my own duplicate — a minimal subset of this change) so there is only one PR to triage.

Diff is unchanged at +71 −1 across 2 files. Happy to split the `_CONTEXT_OVERFLOW_ERROR_CODES` extraction into its own commit if that makes review easier.

anhtahaylove:
CI status at current head `875a0969`: GitHub created a single Actions run ([34561836584](https://github.com/NousResearch/hermes-agent/actions/runs/34561836584)) but it ended immediately as `startup_failure` / `BuildFailed`, with no workflow name, no jobs, and no downloadable logs. This is Actions infrastructure/config startup failure — no test executed and there is no code failure to fix in this PR.

The previous head `39ea3c61` completed all three workflows successfully, and `git merge-tree origin/main pr98148` currently reports no conflicts. Maintainer action needed: rerun/approve CI for head `875a0969`; if GitHub keeps emitting `BuildFailed` before job creation, retry the workflow from the Actions UI rather than changing code.

anhtahaylove:
**Request: one rerun of the startup-failed CI run (no code change needed)**

This PR's only CI run is stuck in `startup_failure`, and I don't have permission to rerun it myself (`gh run rerun` returns `Must have admin rights to Repository`). Asking a maintainer to press rerun.

Evidence that this was infrastructure, not this PR:

**1. The run never started.** Run [`34561836584`](https://github.com/NousResearch/hermes-agent/actions/runs/34561836584) has `conclusion=startup_failure`, an empty workflow name, and **0 jobs**. Nothing in this branch was ever executed, so there is no failing test or lint to fix.

**2. It was part of a repo-wide wave.** Counting all runs in this repository by day:

| Date | Total runs | `startup_failure` |
|---|---:|---:|
| 2026-09-09 | 12,039 | 145 |
| 2026-09-10 | 8,431 | 23 |
| 2026-09-11 | 8,918 | **70** ← this run |
| 2026-09-12 | 8,651 | **0** |
| 2026-09-13 | 1,481 | **0** |

The wave included runs on pushes to `main`, not just fork PRs.

**3. The wave is over.** Zero `startup_failure` runs repo-wide across 2026-09-12 and 2026-09-13. Whatever caused it has already been resolved, but this run won't retry on its own.

**4. This branch was green before.** The same branch passed all three workflows (CI, Nix flake check, Docker Build) at `39ea3c6125` on 2026-09-08. The PR touches two files — `agent/error_classifier.py` (+17/-1) and `tests/agent/test_error_classifier.py` (+54) — and no workflow or `.github/` file.

Current head is `875a09693980f60994fcf8330bf729b7c7725f9b`, with exactly one run against it (`run_attempt=1`). I've deliberately not pushed anything to force a fresh run, since an empty commit would only add noise and reset the review state. Happy to rebase instead if you'd prefer that over a rerun.

If it fails again after a rerun, I'll treat it as a real bug and investigate.

anhtahaylove:
Diagnosis for the `startup_failure` here, since it is the only PR of mine in this state.

Run [34561836584](https://github.com/NousResearch/hermes-agent/actions/runs/34561836584) on `875a096939` has `path=BuildFailed` and an empty workflow `name` — GitHub failed to instantiate a workflow at all, so no job, step, or log exists (the logs endpoint returns 404). That is why this PR shows `0 check-runs` and cannot be fixed by approving a run.

Evidence that the branch content is not the cause:

- The PR touches only `agent/error_classifier.py` and `tests/agent/test_error_classifier.py` — no workflow files.
- `.github/workflows/` on this branch has 35 files, the same count as `main`.
- The two preceding commits on this same branch queued runs normally (`45a6101f36` produced 100 runs, `c3e01c753d` produced 5).
- No other open PR of mine has a `BuildFailed` run; this head has exactly one run and it is that failure.

The branch is **819 commits behind `main`**, so the most likely trigger is a transient instantiation failure against a stale workflow snapshot. A plain re-run of that run id should clear it; if it does not, I can rebase onto current `main` to force a fresh workflow resolution. Say which you prefer and I will do it.

anhtahaylove:
Rebased onto current `main` — the `startup_failure` is gone.

- Head: `875a096939` → **`8c3e4dfb92`**, rebased onto `422bc9bde9` (was 830 commits behind).
- Patch content is unchanged. Comparing the pre- and post-rebase diffs, the only differences are hunk line offsets (`@@ -220` → `@@ -225`, etc.); no added, removed, or altered lines.
- `tests/agent/test_error_classifier.py` on the rebased tree: **158 passed**.

The new head produced three normal workflow runs (`CI`, `Nix flake check`, `Docker Build, Test, and Publish`) instead of the previous `path=BuildFailed` instantiation error — which confirms the failure came from the stale workflow snapshot, not this branch. All three now sit at `action_required`, so this PR has joined the other ten awaiting a fork-run approval; no rerun is needed anymore.

PR is `MERGEABLE`.

PR #98150, opened 2026-08-29
feat(plugins): expose tool progress callbacks to handlers

topolar:
## Summary

- expose the existing per-turn `tool_progress_callback` to registry-backed custom/plugin tool handlers
- preserve callback identity through direct, deferred `tool_call`, concurrent, and both sequential executor paths
- keep the API backward compatible with an optional `None` default and no model-visible schema or transcript changes

## Motivation

Custom long-running tools can currently return only a final result even though the gateway already owns a turn-scoped progress renderer. The registry handler receives task/session context but not the existing presentation callback, so standalone plugins cannot publish safe local progress into the same editable gateway bubble.

This adds only the missing callback plumbing; it does not introduce a new tool, model call, message role, or persistence path.

## Verification

- `48 passed` — model dispatcher, invoke path, and sequential executor tests
- `15 passed` — gateway callback/render seam tests
- `73 passed` — registry and deferred tool-search bridge tests
- `136 passed` total across the focused groups
- `python -m py_compile` on all changed source/tests
- `git diff --check`
- independent review: PASS
- live Telegram E2E with an external role-task plugin showed multiple sanitized child activities updating during a 36-second custom tool invocation

## Safety

- callback defaults to `None`
- no model-visible tool schema changes
- no progress events are added to the model transcript
- no extra model/API calls are made for progress
- existing registry exception handling and middleware semantics are preserved

### DIFF
--- a/agent/agent_runtime_helpers.py
+++ b/agent/agent_runtime_helpers.py
@@ -3588,6 +3588,7 @@ def invoke_tool(agent, function_name: str, function_args: dict, effective_task_i
                 enabled_toolsets=getattr(agent, "enabled_toolsets", None),
                 disabled_toolsets=getattr(agent, "disabled_toolsets", None),
                 tool_request_middleware_trace=list(_tool_middleware_trace),
+                progress_callback=getattr(agent, "tool_progress_callback", None),
             )
             if skip_tool_execution_middleware:
                 dispatch_kwargs["skip_tool_execution_middleware"] = True
--- a/agent/tool_executor.py
+++ b/agent/tool_executor.py
@@ -2534,6 +2534,7 @@ def execute_tool_calls_sequential(agent, assistant_message, messages: list, effe
                             tool_request_middleware_trace=list(middleware_trace),
                             enabled_toolsets=getattr(agent, "enabled_toolsets", None),
                             disabled_toolsets=getattr(agent, "disabled_toolsets", None),
+                            progress_callback=getattr(agent, "tool_progress_callback", None),
                         )
 
                 (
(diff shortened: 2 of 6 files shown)


Enough1122:

**Overall:** Exposes gateway's per-turn `tool_progress_callback` to registry-backed custom/plugin handlers so long-running tools can stream progress instead of only returning final result.

**What it does**
- `model_tools.py:handle_function_call` gains `progress_callback` param, forwards to `registry.dispatch(..., progress_callback=progress_callback)` (both direct and deferred `tool_call` bridge paths, preserving callback identity) and through `_dispatch(next_args)` (`handle_function_call(..., progress_callback=progress_callback)` via sandbox path).
- `agent/tool_executor.py` forwards `getattr(agent, "tool_progress_callback", None)` into both concurrent and sequential executor `_execute` dispatch (`handle_function_call(..., progress_callback=...)`).
- `agent/agent_runtime_helpers.py:invoke_tool` also forwards via `_ra(...).handle_function_call`.
- Tests `test_plugin_progress_callback.py`, `test_tool_call_incremental_persistence.py::test_sequential_dispatch_forwards_plugin_progress_callback`, `test_model_tools.py` (direct, deferred, identity).

**Non-blocking notes**
- `progress_callback` defaults to `None`, `registry.dispatch` presumably handles `None` gracefully — backward compatible, no model-visible schema/transcript change.
- Sequential vs concurrent executor tests both parameterized (`quiet_mode` True/False) — ensures callback not dropped by tool-search deferred path.
- Consider documenting that plugin handlers receive optional `progress_callback` callable and should guard `if progress_callback: progress_callback(text)`; not required for approval.

Clean plumbing.

PR #94874, opened 2026-08-25
fix(cli): persist a launch-context independent desktop-entry Exec

gokhanyildirimlar:
## Bug Description

`resolve_exec_command()` (introduced by #90492) derives the `.desktop` `Exec=` line from `resolve_hermes_bin()`, whose **first priority is `sys.argv[0]`**. But the value of `argv[0]` depends on *how Hermes was started*, and `hermes desktop` rewrites the entry on every launch. So the persisted `Exec=` line is a function of the previous entry itself — a self-referential loop that pins a broken, fragile command line permanently.

Three launch contexts produce three mutually-overwriting Exec lines (verified live on a git install, v0.20.5, Linux). Note the first row: even the "healthy" form degrades at its next rewrite, because the wrapper execs the venv python with the repo script — argv[0] is checkout-internal in *both* entry states:

| Launch context | `argv[0]` | Produced `Exec=` |
|---|---|---|
| Entry says `/home/<u>/.local/bin/hermes desktop` | `~/.local/bin/hermes` (bash wrapper exec'ing venv python + repo script) | `<uv-python> <repo>/hermes desktop` ❌ degrades on next rewrite |
| Entry says `<uv-python> <repo>/hermes desktop` | repo script `hermes-agent/hermes` (`#!/usr/bin/env python3`) | `<uv-python> <repo>/hermes desktop` — pinned again, never self-heals |
| Relaunch with an executable interpreter as argv[0] (programmatic relaunch paths) | interpreter binary | `<interpreter> desktop` ❌ not a runnable command line for a DE |

After any rewrite, both entry states converge on the broken interpreter-pinned form and stay there — #90492's own `_needs_interpreter()` branch sees the repo script's `#!/usr/bin/env python3` shebang and re-emits `sys.executable <script>`: the exact form that created the state. The fix cannot repair installations it was written to fix; it only preserves them.

Two concrete failure modes:

1. **Bootstrap deadlock (reproduced).** Once the entry holds the interpreter-pinned form, no subsequent `hermes desktop` run restores the durable wrapper form. I had to hand-write the correct `Exec=` line to break the cycle.
2. **Fragile interpreter path.** The pinned form hardcodes the uv-managed interpreter (`~/.local/share/uv/python/cpython-3.11.15-.../bin/python3.11`). A `uv python upgrade` removes that directory and the launcher dies with a silent DE failure (`Terminal=false`) — the exact non-durability concern raised in #80547's review thread.

[...]

PR #96777, opened 2026-08-28
fix(desktop): keep venv interpreter unresolved in Linux .desktop Exec (#92095)

i-am-LC:
## What & why

`hermes desktop` installs the XDG launcher entry (`~/.local/share/applications/hermes.desktop`) with an `Exec=` built from `Path(sys.executable).resolve()`. In venvs whose `bin/python` is a **symlink into a versioned runtime tree** (uv-managed CPython, and Hermes' own generation-stamped runtime python under `.hermes-runtime/python/generation-<n>/…`), `.resolve()` dereferences the symlink and writes the **base interpreter** into `Exec=`.

Two consequences — both silent (`Terminal=false` hides the traceback):
1. The base interpreter has no access to the venv's site-packages, so the launch dies on the first third-party import (`ModuleNotFoundError: No module named 'yaml'`).
2. The written path is pinned to a generation directory that the **next `hermes update` deletes**, so the launcher icon stops launching after every update.

The terminal works because `~/.local/bin/hermes` is a wrapper that execs the venv python.

**Fix:** keep `sys.executable` **unresolved** (it is already absolute) in both branches of `resolve_exec_command()` — the interpreter-prefix branch and the `-m hermes_cli.main` module fallback — via a small `_launcher_python()` helper, and make `_needs_interpreter()` recognise venv console-script shebangs by matching against **both** the unresolved and the resolved interpreter dirs. The venv symlink path is stable across updates, and PEP 405 venv detection keys off the path the interpreter is invoked with.

## Verification

- **Regression tests: 4 failed on unpatched `main`** (`test_exec_prefixes_interpreter_for_env_shebang_python_script`, `test_exec_keeps_symlinked_venv_interpreter_unresolved`, `test_exec_module_fallback_keeps_symlinked_venv_interpreter_unresolved`, `test_venv_shebang_matches_unresolved_executable`); **18 passed, 2 skipped** (macOS/Windows no-op marks) with this patch. Python 3.11, pytest 9.1.1.
[...]

PR #95190, opened 2026-08-26
fix(desktop): keep venv interpreter unresolved in .desktop Exec

thehamsti:
## What does this PR do?

Fixes the Linux launcher icon silently failing after every update on uv-based installs (the standard installer layout).

**Root cause:** `resolve_exec_command()` prefixes the resolved `hermes` script with `str(Path(sys.executable).resolve())`. uv-created venvs make `venv/bin/python` a **symlink** into the shared base-interpreter tree (`~/.local/share/uv/python/cpython-3.11.16-.../bin/python3.11`), and `.resolve()` follows that symlink. CPython discovers `pyvenv.cfg` from the executable path *as invoked*, so the generated `Exec=` line ran the **bare base interpreter with no venv site-packages** and died instantly on the first third-party import:

```
ModuleNotFoundError: No module named 'yaml'
```

— silently, because `Terminal=false`. `hermes desktop` from a terminal worked (the bash wrapper execs the venv python), and each launch/rewrite of the entry re-broke it, which is why it recurred after every update.

**Fix:** use `sys.executable` verbatim (it is already absolute). The running process's imports came through exactly that path, so it is always at least as correct as the resolved path — resolving can only lose venv context, never gain anything.

The existing code comment already states the intent ("sys.executable is the interpreter actually running Hermes (the venv one)"); `.resolve()` defeated it.

## Related Issue

Fixes #94058

(Also matches the symptoms in #92882, #92095, #92086, #91504, and duplicates #94564 / #94110. Related prior PR #94593 was closed by its author; this takes the simpler root-cause approach instead of the sibling-venv heuristic, and also fixes the `-m hermes_cli.main` fallback branch which had the same `.resolve()`.)

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `hermes_cli/linux_desktop_entry.py` — `resolve_exec_command()`: use `sys.executable` without `.resolve()` in both the script-prefix branch and the `-m hermes_cli.main` fallback branch; comment explains why resolving is wrong.
- `tests/hermes_cli/test_linux_desktop_entry.py` — updated `test_exec_prefixes_interpreter_for_env_shebang_python_script` to expect the unresolved interpreter; added `test_exec_keeps_venv_symlink_interpreter_unresolved`, a regression test that builds a uv-style layout (venv `bin/python` → base-store symlink) and asserts `Exec=` uses the unresolved venv path and never mentions the base interpreter.

## How to Test

[...]

PR #98154, opened 2026-08-29
fix(desktop): tear down backend on SIGTERM/SIGINT

amitamit10:
## What this does

Adds `SIGTERM`/`SIGINT` handlers to the Electron main process so a signal-based stop tears the backend down before exit.

Only a graceful window-close reached the `before-quit` → `backendShutdown` path. A signal-based stop — session logout, `kill`, a supervising process — bypassed it entirely, because Node's default disposition for `SIGTERM`/`SIGINT` is immediate termination. The backend child and any pooled backends were then reparented to `init` and kept running after the desktop process was gone.

```js
let isExitingFromSignal = false

function handleTerminationSignal() {
  if (isExitingFromSignal) {
    return
  }

  isExitingFromSignal = true
  void exitAfterBackendShutdown(0)
}

process.on('SIGTERM', handleTerminationSignal)
process.on('SIGINT', handleTerminationSignal)
```

The `isExitingFromSignal` guard makes a repeated signal a no-op rather than re-entering teardown.

## Scope change since the original submission

This PR originally also carried a `.desktop` `Exec=` venv fix. **That half has been dropped** — 6fe933e7 (*"fix(cli): launch-context-independent Linux desktop-entry Exec"*, salvaged from #94874) landed upstream with a strictly more general fix: it keeps the lexical `sys.executable` when a `pyvenv.cfg` sits at or above it in the tree, and falls back to the dereferenced path otherwise. That supersedes what this PR did, so on rebase I took upstream's version of `hermes_cli/linux_desktop_entry.py` unchanged.

What remains is only the signal-teardown change, which is still unique — there are currently no `process.on(` signal handlers in `apps/desktop/electron/main.ts` on `main`.

## Verification

Rebased onto current `main` (1 ahead, 0 behind).

- `npm run typecheck` → `tsc -p tsconfig.electron.json --noEmit` passes clean
- `eslint electron/main.ts` passes clean

## Notes

`exitAfterBackendShutdown` is invoked fire-and-forget (`void`), as raised in the earlier automated review on this PR. If `backendShutdown` can hang, the signal handler will not force an exit. Happy to add a bounded timeout / `SIGKILL` escalation here if maintainers prefer that belt-and-braces.

### DIFF
--- a/apps/desktop/electron/main.ts
+++ b/apps/desktop/electron/main.ts
@@ -17648,3 +17648,24 @@ app.on('window-all-closed', () => {
     app.quit()
   }
 })
+
+// SIGTERM/SIGINT (session logout, `kill`, a supervising process stopping us)
+// bypass the normal window-close -> before-quit -> backendShutdown path
+// entirely, since Node's default disposition for those signals is to die
+// immediately. Without an explicit handler here the backend child (and any
+// pooled backends) get reparented to init and keep running after the
+// desktop process is gone. Handle both so a signal-based stop always tears
+// the backend down before the process exits.
+let isExitingFromSignal = false
+
+function handleTerminationSignal() {
+  if (isExitingFromSignal) {
+    return
+  }
+
+  isExitingFromSignal = true
+  void exitAfterBackendShutdown(0)
+}
+
+process.on('SIGTERM', handleTerminationSignal)
+process.on('SIGINT', handleTerminationSignal)


Enough1122:

**Overall:** Two related Desktop reliability fixes for venv + signal termination: `.desktop` Exec now stays inside venv, and SIGTERM/SIGINT always tears down backend/pooled backends before exit.

**What it does**
- `hermes_cli/linux_desktop_entry.py` changes `resolve_exec_command()` and `_needs_interpreter()` to use `Path(sys.executable).parent` / `sys.executable` directly instead of `Path(sys.executable).resolve()`. A uv venv's `bin/python*` is a symlink to base interpreter; `.resolve()` collapsed to base path which has no `pyvenv.cfg` and loses venv `site-packages`, causing silent launch failure (`Terminal=false`). Unresolved path preserves venv detection (CPython keys off unresolved `sys.executable`). `argv` construction now `sys.executable` not `str(Path(...).resolve())`.
- `apps/desktop/electron/main.ts` adds `handleTerminationSignal` (`isExitingFromSignal` guard, `void exitAfterBackendShutdown(0)`) wired to `process.on('SIGTERM')` and `SIGINT`. Without this, signals bypass `window-close → before-quit → backendShutdown` and reparent backend children to `init`.

**Non-blocking notes**
- `sys.executable` may be relative in some invocations; `Path(...).parent` will still correctly compare `exe_dir not in shebang` for venv detection — relative path unlikely via Hermes entry, but worth noting.
- `exitAfterBackendShutdown` is fire-and-forget (`void`) — if it hangs, process still stays alive; ensure `backendShutdown` has its own timeout to avoid stuck signal handler. Existing `isExitingFromSignal` prevents double-entry.
- Comment notes `Terminal=false` silent failure — good context for why `.resolve()` caused invisible breakage.

Minimal, correct.

meviusisback:
Tested this fix on Omarchy (Arch Linux, Hyprland, uv-managed venv) — confirmed it resolves the issue. The Exec line now correctly uses the venv symlink instead of the dereferenced base interpreter.

For context: this bug hits every uv-based install (which is the default installer path). The .desktop file gets rewritten on every [hermes] using password-store backend: gnome-libsecret
[hermes] install stamp: 446563262e54 (main) from local
✓ Desktop packaged app is up to date (content stamp matches)
✓ Desktop launcher entry installed: /home/alberto/.local/share/applications/hermes.desktop
→ Launching packaged Hermes Desktop: /home/alberto/.hermes/hermes-agent/apps/desktop/release/linux-unpacked/Hermes launch, so the breakage recurs after updates even if manually fixed. This blocks launching Hermes from any DE menu, dock, or favorites.

The SIGTERM/SIGINT fix is a nice bonus — I've seen orphaned backend processes after session logouts. Both changes are small and well-scoped.

There are 9+ duplicate PRs for the .desktop Exec issue (#95190, #94115, #96777, #92278, #98279, #94051, #97992, #92790, #92516). Hoping this one gets reviewed as it covers the most ground. The original issue #90292 has been open since Aug 18.

amitamit10:
Hi maintainers — this PR fixes the uv venv .desktop Exec issue (affects all default uv installs) + SIGTERM/SIGINT backend teardown. Tested on Omarchy/Arch (confirmed fix) and includes AI review 'Minimal, correct'. There are ~9 duplicate PRs for the same Exec bug — this one covers the most ground (both Exec + signal handling). Would appreciate a review when you have a moment. Thanks!

amitamit10:
Rebased onto current `main` and **narrowed the scope of this PR.**

6fe933e7 (*"fix(cli): launch-context-independent Linux desktop-entry Exec"*, salvaged from #94874) landed upstream on Aug 31 with a strictly more general version of the `.desktop` `Exec=` venv fix this PR carried — `_running_interpreter()` keeps the lexical `sys.executable` when a `pyvenv.cfg` sits at or above it, and falls back to the dereferenced path otherwise. That covers the uv-symlink case this PR handled plus the bare repo-script/shebang case it did not, so I've **dropped that half** and taken upstream's `hermes_cli/linux_desktop_entry.py` unchanged on rebase.

What's left is just the `SIGTERM`/`SIGINT` backend teardown in `apps/desktop/electron/main.ts` — 21 lines, still unique (`main` currently has no `process.on(` signal handlers at all). Title and description updated to match.

Verified on the rebased branch: `tsc -p tsconfig.electron.json --noEmit` clean, `eslint electron/main.ts` clean.

Re: the earlier automated review note that `exitAfterBackendShutdown` is fire-and-forget (`void`) — that's still true here. If `backendShutdown` can hang, the handler won't force an exit. I left it as-is to keep the diff minimal, but I'm glad to add a bounded timeout with `SIGKILL` escalation if that's preferred.

amitamit10:
Verified this change end-to-end on real hardware — **Zorin OS 18.1 (Ubuntu-based), GNOME/Wayland, Electron 40.10.2**.

Built and ran the packaged app from this branch (install stamp `482393fb198f`), confirmed the Python backend was spawned as a direct child of the Electron main process, then sent `SIGTERM` to the **main process only** — not the process group, so nothing but the handler in this PR could reap the child:

```
before: main=Sl  backend=Sl          # backend is a direct child of Electron main
$ kill -TERM <electron-main-pid>     # main process only
BOTH EXITED after 2s
final:  main=GONE  backend=GONE
```

Without this handler the backend is reparented to `init` and keeps holding its port, because Node's default disposition for `SIGTERM` terminates the main process before `before-quit` can run. With it, `exitAfterBackendShutdown(0)` reaps the child before exit.

Two caveats on what this does and doesn't establish:

- This is the fix-present case only. I did not build `main` to observe the orphan directly, so the "without this" half rests on reading the code path (there are currently no `process.on(` signal handlers on `main`), not on a measurement.
- Teardown completed in ~2s here, so the fire-and-forget `void exitAfterBackendShutdown(0)` never had to be tested against a hung `backendShutdown`. That remains the open question from the earlier review note — the offer to add a bounded timeout with `SIGKILL` escalation still stands if maintainers want it.

amitamit10:
Gentle ping — this has been ready since Sep 1 (rebased, scope narrowed after 6fe933e7, verified end-to-end on Zorin OS 18.1 + independent confirmation on Omarchy/Arch). Happy to address any feedback whenever you get a chance. Thanks!

PR #71447, opened 2026-07-25
fix(desktop): persist custom backend skins across restarts

asimons81:
## What does this PR do?

Fixes #71446: custom backend/runtime skins can survive a Desktop restart even when their registry is populated only after the first paint.

## Root cause

Desktop used the same strict `normalizeSkin()` path for both:

1. **persisted startup values**, which may legitimately reference a backend YAML skin or runtime SDK theme that has not registered yet, and
2. **live selections**, which should reject unknown names immediately.

That caused a valid persisted custom skin to be rewritten to `nous` during the boot window.

The original version of this PR tried to bridge that window with a `known-skins` cache. Review correctly identified two problems with that approach: the cache auto-seeded validity from the value being validated, and it did not cover named-profile/runtime-plugin persistence.

## Fix

- Keep `normalizeSkin()` strict for live `setTheme()` calls.
- Add a separate persisted-value path that preserves any non-empty, non-retired stored skin name until its registry becomes available.
- Continue rendering unresolved persisted names through the existing Nous fallback rather than treating them as a valid live selection.
- Let the existing backend/user/SDK registry reactivity repaint automatically when that late-bound name resolves.
- Apply the same persistence contract through `skinPref`, so it covers the legacy global slot and named profiles without maintaining a second cache of skin names.

This removes the `known-skins` cache entirely and covers backend YAML skins plus runtime Desktop SDK themes with one late-binding contract.

## Regression coverage

Added coverage for:

- unresolved global skin persistence across restart
- unresolved named-profile skin persistence
- late runtime SDK theme registration repainting the persisted theme
- strict rejection of an unknown **live** theme selection
- retired skin migration to the default
- existing mode validation remaining strict

## Manual test

1. Set `display.skin: trt` and create `~/.hermes/skins/trt.yaml`.
2. Apply `trt`, fully quit Desktop, then relaunch.
3. During boot the unresolved name may paint through the Nous fallback briefly.
4. Once the backend skin registers, Desktop should repaint `trt` automatically without requiring `/skin trt` again.
5. Repeat with a named profile/runtime SDK theme to verify the same late-binding behavior.

## Related

Fixes #71446

Thanks to @ildunari for identifying the generic persisted-vs-live split and the named-profile/runtime-plugin gap, and to @teknium1 for catching the self-seeding validation flaw.

PR #98155, opened 2026-08-29
fix(desktop): preserve backend-skin theme across app restarts

cargo8:
## Bug Description

A skin activated from the backend (e.g. a custom YAML skin like `darkside` in `$HERMES_HOME/skins/`, activated via `hermes config set display.skin darkside` or `/skin`) is forgotten by the desktop app on every relaunch — the app repaints with the built-in default (`nous`) and the user has to re-pick it each session. Setting it in `config.yaml` has no lasting effect for the same reason.

## Root Cause

Two correct-looking safety mechanisms interact to erase the preference:

1. **Boot-time normalization runs before the backend connects.** The boot paint and `normalizeSkin()` execute at module scope, before the gateway connects. At that moment the theme registry (`resolveTheme`) contains only built-in desktop themes and user-installed themes — backend skins are not registered into `$backendThemes` until `gateway.ready`/`skin.changed` arrives (`themes/backend-sync.ts`).
2. `normalizeSkin` resolved the stored name against that incomplete registry and silently rewrote any unresolvable name to `DEFAULT_SKIN_NAME` (`nous`), **erasing the stored preference** in localStorage.
3. The gateway's connect-time seed deliberately does **not** re-apply (`ingestBackendSkin` with `apply: false`, per its own docstring: "a fresh connect never stomps the user's persisted desktop theme"). Since the persisted theme was already erased in step 2, nothing ever corrects it.

Net effect: a backend-skin choice can never survive a restart, while built-in desktop themes persist fine.

## Fix

`normalizeSkin` now keeps any non-empty, non-retired name instead of dropping names that don't resolve *yet*:

```ts
const normalizeSkin = (name: string | null): string =>
  !name || RETIRED_SKINS.has(name) ? DEFAULT_SKIN_NAME : name
```

Safety:
- `deriveTheme()` already falls back to the `nous` seed for unresolvable names, so an unknown name paints a complete, valid palette (no broken/garbled paint) until the backend registers the real skin.
- `activeTheme`'s memo depends on `backendThemes`, so the real palette repaints as soon as `ingestBackendSkin` registers it — no snap-back, no missed paint.
- `previewTheme` only previews names that resolve, and is unaffected.
- The `RETIRED_SKINS` guard is preserved, so genuinely removed skins still normalize away.

## How to Verify

1. Create a custom skin: `hermes config set display.skin darkside` (or any custom `~/.hermes/skins/*.yaml`).
2. Open the desktop app, confirm the skin appears in Appearance.
3. Quit and relaunch the desktop app.
4. Before: app repaints with `nous`; the Appearance grid shows the default selected. After this fix: the backend skin is still selected and painted.

## Test Plan

- [x] Manual verification on macOS (arm64): rebuilt the app, confirmed the patched `normalizeSkin` logic in the packaged bundle, picked a backend skin, restarted, theme persisted
- [ ] Existing `context.test.tsx` suite passes (CI)
- [ ] Consider adding a regression case: stored name that doesn't resolve in the registry is preserved, not rewritten

## Risk Assessment

Low — the change only stops discarding not-yet-resolvable names. Worst case for a garbage stored name is painting the `nous` fallback (same visual outcome as before) until it's either registered by the backend or manually changed; previously it was silently rewritten to the same default.

### DIFF
--- a/apps/desktop/src/themes/context.tsx
+++ b/apps/desktop/src/themes/context.tsx
@@ -50,7 +50,17 @@ const resolveMode = (mode: ThemeMode, systemDark = matchesQuery('(prefers-color-
   mode === 'system' ? (systemDark ? 'dark' : 'light') : mode
 
 const normalizeSkin = (name: string | null): string =>
-  name && resolveTheme(name) && !RETIRED_SKINS.has(name) ? name : DEFAULT_SKIN_NAME
+  // LOCAL PATCH (darkside-boot-paint): do NOT drop names that don't resolve in
+  // the registry yet. Backend skins (custom YAML skins from the CLI side, e.g.
+  // `darkside`) only register into $backendThemes when the gateway connects —
+  // which is AFTER the boot paint and after normalizeSkin first runs at
+  // module scope. Rewriting them to DEFAULT_SKIN_NAME here erased the stored
+  // preference on every relaunch, so a backend-skin choice never survived a
+  // restart. deriveTheme() already falls back to the nous seed for unresolvable
+  // names, so an unknown name paints safely until the backend registers it;
+  // `activeTheme`'s memo re-derives when $backendThemes changes. Only truly
+  // retired names are normalized away.
+  !name || RETIRED_SKINS.has(name) ? DEFAULT_SKIN_NAME : name
 
 /**
  * A stored mode, or `system` when there isn't one.


alt-glitch:

Duplicate of #71447: it implements the same late-bound persisted-skin normalization mechanism. #71447 also includes coverage for global and per-profile persistence while keeping live theme selection strict.

sleepless-ted:
when merge?

PR #92526, opened 2026-08-22
fix(cron): retain profile secret scope through delivery

ghosty-11:
## Problem

Cron installs the active profile's secret scope around `run_job`, then resets it before `_deliver_result`. In multiplex mode, credential-bearing delivery adapters therefore run unscoped: scoped credential reads fail closed, and raw process-environment fallback would risk another profile's value.

## Change

Keep the same profile scope active for the full job lifecycle, including delivery and bookkeeping, and reset it from the outer `finally`. Early returns, ordinary exceptions, hard interrupts, and ownership-loss paths all pass through that reset.

Extend the existing regression test to require that execution and delivery observe the same scope and that no scope remains after `run_one_job` returns.

## Verification

- Red before the implementation: delivery observed `current_secret_scope() is None` and `get_secret()` raised `UnscopedSecretError`.
- Focused regression: 1 passed.
- Cron lifecycle selection: 62 passed.
- `ruff check cron/scheduler.py tests/cron/test_run_one_job.py`: passed.

PR #91795, opened 2026-08-21
fix(cron): deliver profile-store jobs with the owning profile's bot under multiplex

vinoth12940:
## What does this PR do?

Fixes cron delivery identity for secondary profiles under a multiplexed gateway (`multiplex_profiles: true`): a job owned by profile X now **delivers through profile X's own bot** instead of the default profile's bot.

Implements the expected behavior from #83182 end-to-end:

1. **Secret scope held through delivery** (`cron/scheduler.py`) — the job-owning profile's secret scope was previously reset in a `finally` around `run_job` *alone*, before `_deliver_result` ran. Delivery then resolved `TELEGRAM_BOT_TOKEN` with no scope active — falling back to `os.environ` (empty or another profile's value in a multiplex unit). The scope is now installed by the `run_one_job` wrapper and reset in the wrapper's `finally`, after the full execute → save → **deliver** → mark sequence (including the failure-path delivery in the body's outer `except`).

2. **Per-profile adapter map for the tick** (`cron/scheduler_provider.py`) — the multiplex ticker passed the *shared* default-profile adapter map to every profile's `cron_tick`. It now resolves each profile's own map via the new `profile_adapters` kwarg (`_profile_tick_adapters` helper): secondary profiles get `_profile_adapters[<name>]`; the default profile — and profiles with no secondary adapters — keep the shared map, so single-profile gateways are unchanged.

3. **Gateway wiring** (`gateway/run.py`) — `start_gateway` now passes `runner._profile_adapters` to the cron provider alongside the shared map (additive kwarg; external providers ignore it).

## Related Issue

Fixes #83182

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✅ Tests (adding or improving test coverage)

## Changes Made

- `cron/scheduler.py` — move the profile secret-scope install/reset from `_run_one_job_body` (around `run_job` alone) to the `run_one_job` wrapper, so the scope spans the entire body including both `_deliver_result` call sites
- `cron/scheduler_provider.py` — new `profile_adapters` kwarg on `InProcessCronScheduler.start` / `_start_multiplex`; `_profile_tick_adapters(profile, shared, profile_adapters)` resolves each tick's adapter map; the per-profile loop uses it
- `gateway/run.py` — pass `runner._profile_adapters` (when populated) as `profile_adapters` in `cron_start_kwargs` under multiplex
[...]

PR #83557, opened 2026-08-11
fix(cron): multiplex delivery uses owning profile's bot token and adapter (#83182)

dongjiang1989:
## Summary

Closes: #83182

Under a multiplex gateway (multiple per-profile Telegram bots etc.), cron jobs owned by a secondary profile deliver to the **wrong bot/chat**. Telegram logs show `Thread 9539 not found ... retrying without message_thread_id` — the thread is dropped because the bot token used at delivery time does not own it.

The root cause has two parts:

1. `cron/scheduler.py::run_one_job` installs the job-owning profile's secret scope, then **resets it in the `finally` block between `run_job` and `_deliver_result`**. By the time delivery runs, `current_secret_scope()` is `None` — `load_gateway_config()` → `_getenv()` falls back to `os.environ` (empty in the multiplex unit) — `TELEGRAM_BOT_TOKEN` resolves to the wrong bot (or nothing).

2. The gateway starts the cron ticker with the **shared** `runner.adapters` dict (the default profile's live adapters). `_deliver_result` → `resolve_delivery_transport(...)` only ever consults that shared dict. It never reaches `Gateway._profile_adapters[<job profile>]`, so even when the token happens to be right, cron delivery cannot reach the secondary profile's live adapter.

This is distinct from existing issues #51853 and #54675 — their fix (PR #59315, commit `0f154e7`) covers only adapter **startup**, not cron **delivery-time** scope reset / shared-adapter usage.

## What does this PR do?

Two fixes, both required to close the bug:

- **Part 1 — Scope lifetime**: Move `reset_secret_scope` from the inner `finally` (after `run_job`) to an outer `finally` (after delivery). The profile's secret scope now stays installed through **both** execution and delivery, so `load_gateway_config()` picks up the right `TELEGRAM_BOT_TOKEN` from the owning profile's `.env`.

- **Part 2 — Per-profile adapters**: `gateway/run.py` builds `profile_adapters_by_home` — a mapping from resolved profile home path to that profile's live adapter map (sourced from `runner._profile_adapters` for secondary profiles, `runner.adapters` for the default). This map is threaded through the cron chain: `scheduler_provider.start` → `_start_multiplex` → `cron_tick` → `run_one_job` → `_deliver_result`. `run_one_job` resolves the owning profile's adapter map via the current hermes home (already set by `_start_multiplex`'s per-profile override), so delivery picks the right bot token.

[...]

PR #73363, opened 2026-07-28
fix(cron): deliver each profile's cron via its own adapter in multiplex

MuhammadUsamaMX:
## Summary
In `multiplex_profiles` mode the in-process cron ticker scopes **execution** per profile (HERMES_HOME, cron store, heartbeat, recovery) but delivered **every** profile's cron output through the **default** profile's platform adapter. So a secondary profile's scheduled-job replies / failure summaries were sent by the *default* bot instead of that profile's own bot (e.g. one Telegram bot per profile).

## Root cause
`cron/scheduler_provider.py::InProcessCronScheduler._start_multiplex` iterates each served profile and calls `cron_tick` with the single default-profile `adapters` set for all profiles:
```python
for entry in profile_homes:
    with use_cron_store(home):
        cron_tick(adapters=adapters, ...)   # same default adapters for every profile
```
`adapters` originates from `runner.adapters` (the default/primary profile) in `gateway/run.py`, while each secondary profile's real adapters live in `GatewayRunner._profile_adapters[name]`. Execution was correctly scoped by a61183b, but delivery was not.

## Fix
Thread a `profile_adapters` map (profile name -> adapters) from the gateway into `InProcessCronScheduler.start`, and select the per-profile adapter set in the multiplex tick loop, falling back to the shared `adapters` for the default profile (which owns them). Single-profile path is unchanged; the new parameter defaults to `None`.

## Verification
- Gateway starts healthy; every per-profile Telegram bot connects.
- A secondary profile's cron now delivers via that profile's own adapter/bot instead of the default bot.
- Non-multiplex behavior unchanged (`profile_adapters=None`).

Builds on a61183b (per-profile HERMES_HOME scoping) - same intent, extended from execution to delivery.

## Related issues
- Addresses #70849 — its Problem section is exactly this bug: multiplex cron delivery always uses the default/active profile's `self.adapters`, ignoring `self._profile_adapters`, so a secondary profile's job is delivered by the wrong bot. This PR routes each job's delivery through its owning profile's adapter. The per-job `deliver_profile` override proposed there could layer on top of this default-correct behavior.
- Related: #70620 — the inbound-reply analog (replies sent via the wrong bot adapter under multiplex); different code path (message handler), not touched here.
- Builds on the ticker fix for #69377 (a61183b), which scoped cron *execution* per profile; this extends the same intent to *delivery*.

PR #98156, opened 2026-08-29
fix(cron): deliver multiplex profile jobs from the owning profile's bot

### LINKED ISSUE #83182: Cron delivery uses wrong bot/chat under multiplex (secret-scope reset before delivery + shared-adapter dict)
## Summary
Cron job delivery routes to the wrong bot/chat under gateway multiplexing because the job-owning profile's secret scope is reset **before** delivery, and delivery only consults the shared (default-profile) adapters dict rather than the job profile's live adapters.

## Description
Since the multiplex gateway cutover, a cron job owned by a secondary profile delivers to the wrong topic/bot. Telegram logs show `Thread 9539 not found ... retrying without message_thread_id` — the thread is dropped because the bot token used at delivery time does not own it.

## Root cause
`cron/scheduler.py::run_one_job`:

- Installs the job-owning profile's secret scope: `set_secret_scope(build_profile_secret_scope(_get_hermes_home()))`
- Resets it in the `finally` of the run-job try block: `reset_secret_scope(_scope_token)`
- Then, in a **separate** later try block, calls `_deliver_result(...)` — which runs `load_gateway_config()` and reads `TELEGRAM_BOT_TOKEN` via `_getenv` (gateway/config.py). Because the scope was already reset, `current_secret_scope()` is `None`, so `_getenv` falls back to `os.environ` — empty in the multiplex unit. The owning profile's `.env` is ignored at delivery time.

Additionally, the gateway starts the cron ticker with the **shared** adapters dict (`gateway/run.py` passes `{"adapters": runner.adapters, ...}`), and `_deliver_result` → `resolve_delivery_transport(...)` / `DeliveryRouter(config, adapters)` only ever use that shared dict. It never consults `Gateway._profile_adapters[<job profile>]`, so even with the correct token, cron delivery cannot reach a secondary profile's live adapter.

## Expected behavior
- The job-owning profile's secret scope stays installed through `_deliver_result` and is torn down after delivery completes.
- Cron delivery selects the owning profile's live adapter map (`_profile_adapters[<job profile>]`) when multiplexing is active, falling back to the shared dict for the default profile.

## Environment
- hermes-agent v0.20.0 (2026.8.3), commit 615e69fde
- Multiplex gateway with multiple per-profile crons

## Distinct from existing issues
This is NOT #51853 ("Cron from restricted profiles bypasses tool restrictions and delivers via wrong bot", closed implemented-on-main 2026-07-14) nor #54675 (multiplex adapter-startup token bleed). Their fix (PR #59315, commit 0f154e780) covers only the adapter-**startup** path, not cron **delivery-time** scope reset / shared-adapter usage. This is a new, separate defect.

cargo8:
## Bug Description

Under gateway multiplex (multiple served profiles), cron jobs belonging to a named profile were delivered from the **wrong bot** — typically the default profile's bot token — and in some configurations failed outright with `UnscopedSecretError` before model selection, breaking every profile-owned cron job.

Fixes #83182

## Root Cause

Two independent scope leaks, both fixed:

**1. Secret scope lifetime (`cron/scheduler.py`)**

`_run_one_job_body` reset the profile secret scope in a `finally` right after `run_job` returned. But the delivery step (`_deliver_result`) runs *after* that and resolves platform credentials (`TELEGRAM_BOT_TOKEN`, etc.) through the secret scope. With no scope active, resolution fell back to `os.environ` of whichever multiplex process won the tick — i.e. the wrong bot.

Additionally, cron fires from the ticker thread where no per-turn scope is installed, and `get_secret()` fails closed outside a scope once profile isolation is in play — so `resolve_runtime_provider()` raised `UnscopedSecretError` before model selection, breaking the job entirely.

**2. Adapter maps (`cron/scheduler_provider.py`, `gateway/run.py`)**

The multiplex cron tick passed the **shared default adapter map** to every profile's tick, so delivery rode the default profile's bot even for profile-owned jobs, even when the secret scope resolved correctly.

## Fix

1. The profile secret scope now spans the entire **execute → save → deliver → mark** sequence. The reset lives in the outer `finally` after every delivery call site (failure-alert deliveries included), mirroring the per-turn pattern in `gateway/run.py` (`_profile_runtime_scope`).
2. `gateway/run.py` hands the scheduler a per-profile adapter map (`runner._profile_adapters`) when multiplex is on, and `InProcessCronScheduler` tick selects the owning profile's live adapter map for its jobs. Falls back to the shared map for the default profile / absent maps (single-profile behavior unchanged).

## How to Verify

1. Run a gateway with multiplex profiles (e.g. `default` + a named profile, each with its own bot token).
2. Create a cron job in the named profile that posts to a messaging platform.
3. Before: the message arrives from the default profile's bot (or the job fails with `UnscopedSecretError`).
4. After this fix: the message arrives from the named profile's own bot.

## Test Plan

- [x] `tests/tools/test_cronjob_tools.py`, `test_cron_approval_mode.py`: 101 passed
- [x] Manual verification on a multiplex setup: profile-owned jobs deliver via the owning profile's bot token; default-profile jobs unchanged
- [ ] CI full suite

## Risk Assessment

Low — the secret scope now covers strictly more of the job lifecycle (widening a scope window within a single job body, reset still guaranteed in `finally`). Adapter-map selection only activates under multiplex with per-profile maps supplied; single-profile and default-profile paths are byte-for-byte unchanged.

### DIFF
--- a/cron/scheduler.py
+++ b/cron/scheduler.py
@@ -2301,10 +2301,30 @@ def _get_config_home_channel(platform_name: str):
 
 
 def _env_home_target_chat_id(platform_name: str) -> str:
-    """Return the home chat id from the legacy env mirror only (no config)."""
+    """Return the home chat id from the legacy env mirror only (no config).
+
+    Reads through ``get_secret`` (not raw ``os.getenv``) so a profile-scoped
+    secret scope wins in a multiplex gateway. ``DISCORD_HOME_CHANNEL`` lives in
+    each profile's ``.env``; in a multiplex process the winning cron tick runs
+    with the job-owning profile's scope installed (run_one_job sets it), so
+    reading via ``get_secret`` resolves the OWNING profile's chat id rather
+    than the host process's ``os.environ`` (#83182, chat-id leg — the token
+    leg was fixed earlier; chat id / thread id resolve through the same leak).
+    """
     env_var = _resolve_home_env_var(platform_name)
     if not env_var:
         return ""
+    try:
+        from agent.secret_scope import get_secret
+    except Exception:
+        get_secret = None  # type: ignore
+    if get_secret is not None:
+        value = get_secret(env_var, "")
+        if not value:
+            legacy = _LEGACY_HOME_TARGET_ENV_VARS.get(env_var)
+            if legacy:
+                value = get_secret(legacy, "")
+        return value or ""
     value = os.getenv(env_var, "")
     if not value:
         legacy = _LEGACY_HOME_TARGET_ENV_VARS.get(env_var)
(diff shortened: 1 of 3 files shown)


alt-glitch:

Related: #91795 and #83557 are open competing repairs for #83182. This PR retains the profile secret scope through the scheduler body and selects per-profile adapter maps; maintainers should choose the canonical approach.

teknium1:
Partially merged via #99375 (merged, commit 9eb832aad7): your commit resolving home chat-id/thread-id via `get_secret` under the owning profile's scope (d0b2415afb) was cherry-picked with your authorship — nobody else fixed that leg. The scope-span/profile_adapters commit wasn't carried: earlier submitters @ghosty93 (#92526) and @MuhammadUsamaMX (#73363) covered those legs with tests, and the profile_adapters version here had an unsafe fall-back-to-shared-map default for unconnected secondaries (a named profile without its own bot must fail delivery loudly, not borrow the default's). Thanks for the solid A3 fix!

PR #98157, opened 2026-08-29
fix(desktop): avoid waking Bots for avatar hydration

### LINKED ISSUE #98123: Desktop Bot avatar hydration can start every dormant local profile backend
## Summary

Hermes Desktop Bot Mode can materialize dormant local profile backends during passive avatar hydration, even when the user has not opened, messaged, hovered, or otherwise interacted with those Bots.

In a Windows Desktop setup with 57 non-default local specialist profiles, all 57 specialists had `avatar.png`. On clean startup, the Desktop remained at zero secondary profile backends for the first ~2 minutes, then background Bot UI work gradually activated the entire specialist fleet. The resulting process pattern was 57 unique specialist profiles × 2 Python process layers = 114 secondary process rows.

This is independent of the hidden-session reconciliation fan-out reported in #94872.

## Environment

- Hermes Agent 0.20.6
- Windows 11
- Packaged Hermes Desktop
- 57 non-default local specialist profiles
- 57/57 profiles had an avatar asset

## Reproduction

1. Create a local Bot roster with many dormant local profiles.
2. Ensure each profile has an avatar asset (`avatar.png`).
3. Fully quit Desktop and confirm no specialist profile backends are running.
4. Launch Desktop normally.
5. Do not open a specialist, call `message_agent`, or deliberately prewarm a profile.
6. Observe profile backend processes over several minutes.

Observed in the affected build:

```text
t=0s     secondary profile processes = 0
t=30s    secondary profile processes = 0
t=60s    secondary profile processes = 0
t=120s   secondary profile processes = 0
t=240s   secondary profile processes = 114   # 57 unique profiles × 2 Python layers
```

The full fleet remained/reappeared after subsequent lifecycle churn.

## Root cause

The passive avatar hydration path used source-scoped per-Bot RPC for local roster rows:

```text
passive Bot roster/avatar hydration
→ profiles.get_asset
→ requestForBot(bot, ...)
→ host.requestProfile(...)
→ getConnectionFor(...)
→ ensureRegistryBackend(...)
→ spawn dormant local profile backend
```

The roster rows were source-scoped, so a read-only presentation concern (`profiles.get_asset`) crossed the profile backend lifecycle boundary.

A static audit also showed that `pullServerAvatars()` runs as passive roster work and can invoke this path for avatars that are not already cached locally.

## Local fix / validation

I changed passive avatar reads to use source-primary asset access rather than target-profile activation, while preserving target-profile correctness in the asset lookup itself.

Regression results:

```text
PASSIVE_AVATAR_REQUESTPROFILE_CALLS = 0
FULL_57_ROSTER_PROFILE_WAKE_CALLS   = 0
```

After rebuilding Desktop, a clean 720-second lifecycle run showed:

```text
secondary specialist backends = 0
```

for the entire idle window until unrelated explicit/pointer activation paths were exercised.

The fix preserved interactive specialist activation and Bot Mode behavior.

## Expected behavior

Passive avatar rendering must never require a dormant specialist runtime.

A roster with 200 defined local specialists should be able to render Bot metadata and avatars with approximately zero specialist backend processes until a specialist is explicitly used.

## Suggested acceptance criteria

- Render a full local Bot roster with avatars for 100–200 dormant profiles.
- Assert `requestProfile` / `getConnectionFor` is not used merely to hydrate avatar assets.
- Assert zero dormant profile backends are spawned by avatar hydration.
- Preserve correct alias / target-profile asset lookup semantics.
- Preserve explicit open/message/delegation lazy activation.
- Repeat after primary gateway reconnect and ensure avatar refresh still causes zero fleet activation.

## Relationship to #94872

#94872 covers a separate passive full-roster activation path caused by hidden-session reconciliation calling `session.list` per Bot. This report is a second independent lifecycle leak: a read-only avatar fetch can also materialize each dormant local profile backend.

Fixing only the reconciliation sweep does not close this path.

sylvesterkaczmarek:
## Summary

- keep passive Bot avatar reads and backfills on an already-active source gateway
- never open a dormant target-profile backend just to hydrate roster artwork
- preserve target-profile alias rewriting for asset paths
- leave all non-asset Bot RPCs on the existing strict per-profile route

## Root cause

`pullServerAvatars()` and the avatar backfill path both call `requestForBot()` for source-scoped roster rows.

`requestForBot()` normally routes through `host.requestProfile()`. That is correct for actual Bot work, but it also means a read-only `profiles.get_asset` can materialize the target profile backend. On a large local roster, passive avatar hydration therefore wakes every dormant specialist.

Profile asset handlers do not require the target runtime. They accept the target profile name and read/write that profile's asset path directly.

The router now treats `profiles.get_asset` and `profiles.set_asset` as passive asset operations. When the owning source is already active, they use that source's existing gateway with the target profile preserved in the RPC parameters. When the source is inactive, the best-effort avatar refresh is deferred instead of dialing a profile backend; the existing roster refresh path retries later.

All other Bot methods keep the existing `requestProfile` behavior.

## Testing

Focused regression coverage verifies:

- an active source avatar read uses `host.request` and never `requestProfile`
- an inactive source is deferred without dialing anything
- normal profile RPCs still use strict per-profile routing

Tests not run locally in this environment.

Closes #98123

### DIFF
--- a/apps/desktop/src/plugins/hermes-bots/routing-passive-assets.test.ts
+++ b/apps/desktop/src/plugins/hermes-bots/routing-passive-assets.test.ts
@@ -1,0 +1,105 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import type { RosterRow } from './types'
+
+const { hostMock } = vi.hoisted(() => ({
+  hostMock: {
+    activeConnectionId: vi.fn(() => 'local'),
+    request: vi.fn(),
+    requestProfile: vi.fn(),
+    state: { connectionId: { get: vi.fn(() => 'local') } }
+  }
+}))
+
+vi.mock('@hermes/plugin-sdk', () => ({ host: hostMock }))
+
+const { requestForBot } = await import('./routing')
+
+beforeEach(() => {
+  vi.clearAllMocks()
+  hostMock.activeConnectionId.mockReturnValue('local')
+  hostMock.state.connectionId.get.mockReturnValue('local')
+})
+
+describe('passive Bot asset routing', () => {
+  it('reads an avatar through the already-active source without activating the Bot profile', async () => {
+    hostMock.request.mockResolvedValue({ found: true, data: 'data:image/png;base64,avatar' })
+
+    const bot = {
+      connectionId: 'local',
+      connectionKind: 'local',
+      name: 'designer',
+      route: {
+        connectionId: 'local',
+        mode: 'local',
+        profile: 'designer',
+        targetProfile: 'designer-backend'
+      },
+      sourceScoped: true,
+      targetProfile: 'designer-backend'
+    } as RosterRow
+
+    await expect(
+      requestForBot(bot, 'profiles.get_asset', { name: 'designer', asset: 'avatar' })
+    ).resolves.toMatchObject({ found: true })
+
+    expect(hostMock.request).toHaveBeenCalledWith('profiles.get_asset', {
+      name: 'designer-backend',
+      asset: 'avatar'
+    })
+    expect(hostMock.requestProfile).not.toHaveBeenCalled()
+  })
+
+  it('defers an inactive source instead of dialing its dormant Bot profile', async () => {
+    hostMock.state.connectionId.get.mockReturnValue('source-a')
+    hostMock.activeConnectionId.mockReturnValue('source-a')
+
+    const bot = {
+      connectionId: 'source-b',
+      connectionKind: 'remote',
+      name: 'researcher',
+      remoteSource: true,
+      route: {
+        connectionId: 'source-b',
+        mode: 'remote',
+        profile: 'researcher',
+        targetProfile: 'backend-researcher'
+      },
+      sourceScoped: true,
+      targetProfile: 'backend-researcher'
+    } as RosterRow
+
+    await expect(
+      requestForBot(bot, 'profiles.get_asset', { name: 'researcher', asset: 'avatar' })
+    ).rejects.toThrow(/deferred until source source-b is active/)
+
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Sound fix for passive avatar backfill waking dormant Bot backends. Routing stays strict for non-asset RPCs and correctly fences passive reads via active-source check.

**What it does**
- `apps/desktop/src/plugins/hermes-bots/routing.ts:122` adds `PASSIVE_PROFILE_ASSET_METHODS` (`profiles.get_asset`/`set_asset`) and special-cases `requestForBot` to read via `host.request` on the already-active connection (`host.state.connectionId` / `host.activeConnectionId`) with `scopedBotParams`, never materializing a dormant profile backend.
- If `activeConnectionId !== route.connectionId`, throws `Passive <method> deferred until source <id> is active` instead of dialing `host.requestProfile`; caller (avatar backfill/roster) retries on next active refresh.
- Non-asset Bot RPCs unchanged: still go through `host.requestProfile(route, ...)` (`routing.ts:161`).
- Adds `routing-passive-assets.test.ts` covering active-source success (alias-rewritten `designer-backend`), deferred inactive-source, and strict non-asset routing.

**Non-blocking notes**
- Deferred path throws a plain `Error` that surfaces to the pull/backfill caller — ensure that caller treats `deferred until source ... is active` as silent retryable (best-effort) rather than user-visible failure/toast.
- `host.request` path bypasses per-profile descriptor routing by design (any open gateway can serve asset path) — alias rewriting via `scopedBotParams` is preserved, but verify no profile-scoped auth check is needed on the remote gateway for that asset namespace.
- `activeConnectionId` fallback `'local'` when both getters missing — covers tests but confirm production always exposes one of the two getters; otherwise a headless pass could misroute.

PR #98158, opened 2026-08-29
feat(desktop): add no-payload watchdog for stuck thinking spinner

marginmotor:
## Problem

When a turn goes live (`message.start`) but the backend dies before producing ANY assistant payload, `message.complete` never arrives. The gateway stops heartbeating, so the existing `session.info` running=false settle path never fires. The 5-minute session watchdog (`SESSION_WATCHDOG_TIMEOUT_MS`) was the only backstop — leaving users staring at a thinking spinner for 3-5 minutes.

Affected models: DeepSeek, GLM Flash, and any model that hits `max_iterations` where the summary call hangs or fails silently.

Sessions from bug reports:
- `20260827_154335_e3ed79`
- `20260731_183224_ce6b1a`

## Root Cause

In `use-message-stream/index.ts`, the `completeAssistantMessage` callback only fires on `message.complete` events. When the backend dies mid-turn with no assistant text, that event never arrives. The existing pre-start grace gate (`PRE_TURN_LIVE_SETTLE_GRACE_MS = 15s`) only covers turns that never went live. Once `turnLive=true` and `sawAssistantPayload=false`, nothing settles the turn until the 5-min watchdog.

## Fix

Added a client-side **no-payload watchdog** in `useMessageStream`:

- Scans the per-runtime state map every 5 seconds
- Force-settles any session where:
  - `turnLive=true` (backend confirmed the turn started)
  - `sawAssistantPayload=false` (no assistant text ever arrived)
  - `busy=true` / `awaitingResponse=true`
  - `turnStartedAt` is older than `NO_PAYLOAD_WATCHDOG_MS` (60 seconds)
- Under the updater, re-checks all guards to avoid racing with a late payload
- Triggers `scheduleSessionsRefresh()` so the sidebar clears the working dot

## Files Changed

- `apps/desktop/src/app/session/hooks/use-message-stream/index.ts` — added `NO_PAYLOAD_WATCHDOG_MS` constant + the watchdog `useEffect`
- `apps/desktop/src/app/session/hooks/use-message-stream/no-payload-watchdog.test.tsx` — three test cases (fires after 60s, skips if payload arrived, skips before deadline)

## Testing

Three new test cases:
1. Force-settles a live turn with no payload after the watchdog fires
2. Does NOT settle a live turn that has produced payload
3. Does NOT settle a live turn before the watchdog deadline

Co-authored-by: realkdc <realkdc@users.noreply.github.com>

### DIFF
--- a/apps/desktop/src/app/session/hooks/use-message-stream/index.ts
+++ b/apps/desktop/src/app/session/hooks/use-message-stream/index.ts
@@ -35,6 +35,14 @@ import type { ClientSessionState } from '../../../types'
 import { useGatewayEventHandler } from './gateway-event'
 import { completionErrorText, delegateTaskPayloads, MAX_STREAM_FLUSH_GAP_MS, STREAM_DELTA_FLUSH_MS } from './utils'
 
+// When a turn goes live (message.start) but the backend dies before producing
+// ANY assistant payload, no message.complete ever arrives and the gateway may
+// stop heartbeating entirely. The existing 15s pre-start grace gate only covers
+// turns that never went live; once turnLive=true, a dead backend leaves the
+// spinner pinned until the 5-min session watchdog fires. This constant bounds
+// that wait: if a live turn has produced nothing for this long, force-settle.
+export const NO_PAYLOAD_WATCHDOG_MS = 60_000
+
 interface MessageStreamOptions {
   activeGatewayProfile?: string
   activeSessionIdRef: MutableRefObject<string | null>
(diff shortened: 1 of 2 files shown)


marginmotor:
## Update: rewritten + fully verified

First version used a `setInterval` scan which added a permanent window timer — broke 3 timer-count assertions in `delta-flush.test.tsx` (caught running the full suite after install; the initial push had unverified tests, my mistake).

**Rewritten (a9be914):**
- Lazy-armed: a single bounded `setTimeout` exists only while a session sits in the stuck shape, armed from the `updateSessionState` chokepoint (every stuck-shape write flows through it: `message.start`, `session.info running=true`, optimistic submit arm)
- `fire()` re-checks all guards at fire time AND under the settle updater, so late payloads / normal settles / Stop always win
- Rearms once if a candidate is still inside the grace window; zero idle timers otherwise

**Verification (all green):**
- `tsc --noEmit`: clean
- `use-message-stream/`: **142/142** (incl. the 3 previously broken delta-flush specs)
- session hooks + store + liveness-policy: **2080/2080**

New watchdog specs in `no-payload-watchdog.test.tsx`: force-settles after deadline, spares turns with payload, spares healthy turns pre-deadline, releases the timer after a normal settle.

PR #23804, opened 2026-05-11
feat(gateway): add opt-in AFK follow-up turns

tymrtn:
## Summary

- Adds opt-in gateway-native AFK follow-up scheduling via `gateway.afk_followup`.
- Injects AFK prompts as internal virtual `MessageEvent`s at configured idle thresholds.
- Skips sessions that are currently running or have queued/pending user work.
- Fixes internal synthetic response delivery so an AFK summary is not suppressed just because a later user message is pending.

Fixes #23803

## Root cause

`BasePlatformAdapter._process_message_background` suppresses responses when an interrupt event is set and a pending message exists. That is correct for normal stale user-originated responses, but internal synthetic events are different: the internal AFK summary is the response to an already-started virtual turn and should be delivered before the pending user message drains.

Also, AFK follow-up scheduling itself was not wired into upstream `GatewayRunner`, so configs could contain `gateway.afk_followup` while the runtime had no implementation.

## Test plan

```bash
/Users/tylermartin/.hermes/hermes-agent/venv/bin/python -m pytest \
  tests/gateway/test_afk_followup.py \
  tests/gateway/test_pending_drain_race.py \
  tests/gateway/test_session_race_guard.py \
  tests/gateway/test_pending_event_none.py \
  -q
# 33 passed

/Users/tylermartin/.hermes/hermes-agent/venv/bin/python -m py_compile \
  gateway/afk_followup.py \
  gateway/run.py \
  gateway/platforms/base.py
```

## Safety notes

- AFK follow-up is disabled by default.
- The prompt explicitly forbids destructive, credential-changing, deployment, publishing, message-sending, financial, and irreversible actions unless the exact action was already approved.
- Scheduler uses the existing gateway session store and adapter delivery path; it does not create cron jobs or external side effects itself.

PR #21383, opened 2026-05-07
feat: cross-bot user-level ephemeral status

tymrtn:
Closes #21122.

## Summary

Adds a small cross-bot, user-level ephemeral status layer so multiple gateway profiles (CLI, Telegram, Discord, etc.) can share a single, lightweight view of "what the user is currently doing/where/on what device" without mutating the cached system prompt or relying on long-lived memory.

The change is split into three focused layers:

### 1. Storage — `agent/user_status.py`
- Single shared JSON file at `~/.hermes/state/user_status.json`.
- Atomic write via temp-file + rename, file-lock-friendly, schema-validated.
- Field set matches the issue spec: `device_mode`, `afk_status`, `focus_project`, `quiet_hours_until`, `location`, plus per-field `*_updated_at` timestamps.
- `load()` / `save()` / `update()` helpers with defensive fallbacks (missing/corrupt file → empty status, never raises into the agent loop).

### 2. Mutation — `tools/user_status_tool.py` + `toolsets.py`
- New `user_status` tool registered in `_HERMES_CORE_TOOLS` so it's available across all gateway profiles by default.
- Supports `get`, `set`, and `clear` actions with per-field validation.
- Updates only the fields the caller passes; leaves others untouched and refreshes only those fields' timestamps.

### 3. Injection — `run_agent.py`
- New `_user_status_turn_context()` helper builds a short ephemeral turn-context line from the stored status.
- Hooked into the existing `pre_llm_call` plugin-context aggregation so it rides alongside other per-turn context (read-only; never mutates `_cached_system_prompt`, preserves prompt caching).
- Stale fields produce softened phrasing ("possibly …", "may be …") instead of hard claims so the model doesn't act on outdated state with full confidence.

### Staleness rules (match issue spec)
- `device_mode` → 12h
- `afk_status` → 30min
- `focus_project` → 24h
- `location` → 24h
- `quiet_hours_until` → compared against its own timestamp value vs. now

## Tests

36 new tests, all green:

- `tests/agent/test_user_status.py` — 14 storage tests (load/save/update, atomic write, schema validation, missing-file and corrupt-file fallbacks).
- `tests/tools/test_user_status_tool.py` — 16 tool tests (get/set/clear actions, per-field validation, partial updates, timestamp behavior).
- `tests/test_user_status_injection.py` — 6 injection tests (per-field rendering, staleness softening, empty/missing status no-ops, exception safety in the turn-context path).

```
[...]

Issue #21122, opened 2026-05-07
Cross-bot user-level ephemeral status (device mode, AFK, focus project, quiet hours)


tymrtn:
## Problem

Hermes provides two surfaces for non-conversational context:
- **`USER.md`** for durable user facts (cached as part of the system prompt)
- **`_plugin_turn_context`** in `run_conversation()` for per-turn dynamic content

There's a gap between them: **frequently-flipping user state that should be visible to every profile** — things like:

- Device/form-factor mode (e.g. mobile vs desktop, where reply length and attachment behavior should differ)
- Availability state (available / short break / long break)
- Current focus project (so unrelated profiles can defer proactive nudges)
- Quiet hours
- Travel/location-dependent defaults

Storing these in `USER.md` per-profile creates three issues:

1. Every profile needs updating when the flag changes (N writes per flip)
2. Each profile pays a system-prompt cache invalidation per edit
3. Profiles can drift out of sync if any one write fails or lags

The architectural pattern in `software-development/hermes-ephemeral-turn-context` already covers the right shape (inject per-turn, don't bake into the cached prefix). This issue proposes a small concrete implementation of that pattern for cross-profile user state.

## Proposed approach

**Single shared status file + per-turn ephemeral injection.**

### Storage
A JSON file under `~/.hermes/state/` containing the active flags, last-updated timestamps per field, and an attribution field for the last writer. Atomic read-modify-write so concurrent profiles don't clobber each other.

### Injection
A helper on `AIAgent` invoked from the `run_conversation()` ephemeral path that:
- Reads the file
- Applies per-field staleness rules
- Returns a short status line for the active fields
- Appends to `_plugin_turn_context` only — never to `_cached_system_prompt`

### Mutation
A small tool (e.g. `user_status`) registered in the gateway toolset with `get` / `set` operations, available to any profile that wants to update the shared state in response to a user signal.

### Optional
Slash commands for fast manual switching (`/status` family).

## Why not extend user memory

User memory is in the cached prefix. Any edit invalidates the cache for the rest of the session, which is the wrong cost profile for state that may flip several times a day.

## Why a single shared file vs per-profile state

A shared file scales to N profiles at the cost of one write and one read per turn. Per-profile copies require either a broadcast mechanism on every change or accept eventual-consistency drift — neither is ideal for state the user expects to be globally true.

## Sketch of acceptance criteria

- `~/.hermes/state/` created on first use
- Status JSON schema with atomic-write helper
[...]

PR #98162, opened 2026-08-29
feat(gateway): add durable global AFK availability

lorenzespinosa:
## Summary

- Adds a durable, global AFK availability state (`/afk [on [reason] | off | status]`, plus `!afk` for Slack threads where native slashes are blocked). State persists across restarts in a dedicated file written `0600` via no-follow temp file + atomic rename. On POSIX, both the file and containing directory entry are fsynced before success is reported; symlink leaves are rejected. Windows retains atomic/read-back semantics without claiming POSIX directory durability.
- While AFK is engaged, **approval prompts fail closed on every interactive surface**: new approval requests are denied before their notification callback runs, approvals already pending in messaging/API/TUI/ACP queues are denied and signaled so tool threads unblock immediately, and stale button/API responses arriving during AFK cannot grant once/session/always approval. The agent receives an automatic-AFK explanation, not the false claim that the user denied it. Absence is never consent.
- Turns on every conversation surface (CLI, API, TUI/desktop, ACP, cron, messaging) are told the user is away — status only. String turns use the API-bound `api_content` sidecar so historical bytes stay prompt-cache stable; multimodal turns append the note only to a deep-cloned outgoing API list. System prompt and role/history structure are untouched, and nothing is added while the user is available.
- The operator's reason (`/afk on picking up the kids`) is shown back in `/afk status` and stored bounded and neutralized — it is never included in the note sent to the model.
- `/afk` is registered with `busy_policy="dispatch"` and wired into the shared plain-command handlers so stepping away *while the agent is working* (the common case) reaches its handler mid-run; also available through the TUI slash worker. On Slack it is `/hermes afk`-only (not a native slash): Slack blocks native slashes inside threads, which is exactly where someone announces they're stepping away, and a native `/afk` would sit confusingly next to Slack's built-in `/away`.

### #21383 / #21122

#21383 predates this PR and introduced the broader cross-bot user-status design. This PR should be the canonical owner of the machine-global AFK availability and approval-gating truth. It supersedes only #21383's independently writable `afk_status` slice while preserving Tyler Martin's credit and the complementary `device_mode`, `focus_project`, `quiet_hours_until`, and `location` fields. If both land, #21383 should derive availability from `agent.afk` or drop its second AFK store rather than allow contradictory truths.

### #23804

Complementary. #23804 schedules opt-in follow-up *turns* while sessions are idle; this PR adds explicit operator availability that gates approvals fail-closed and annotates outbound turns across all surfaces. Both touch `gateway/run.py`, so whichever lands second must rebase and rerun the affected gateway/AFK tests.

### #98011

Complementary Telegram master-task notification work. It touches `tools/approval.py` and `gateway/run.py`; whichever lands second must rebase and re-prove the approval payload and notification path rather than overwrite either ownership change.

## Review remediation

Review `5059279332` was addressed at exact head `7a00d63e09077b536c262d70e750920302ec30f3`:

- POSIX `engage()` and `clear()` now fsync the parent directory after `replace`/`unlink`, including privacy-failure cleanup, and propagate sync failures instead of reporting success.
- `pre_approval_request` hooks and notifier callbacks now run outside the machine-global AFK transaction. Admission remains locked; the same request ID must still be pending before notification; AFK drains/revokes pending entries; every grant retains final availability revalidation.
- Torture coverage proves a blocked hook or notifier cannot delay AFK commit and a late stale response cannot grant `once`, `session`, or `always` authority.
- Independent exact-SHA review found no blocking issues; two minor fail-closed status-message mismatches remain.

## Test plan

```bash
python -m pytest \
  tests/agent/test_afk_state.py \
  tests/agent/test_afk_turn_note.py \
  tests/gateway/test_afk_command.py \
  tests/gateway/test_afk_approval_gate.py \
  tests/gateway/test_afk_slack_bang.py \
  tests/gateway/test_slash_access_dispatch.py \
  tests/agent/test_api_content_sidecar.py \
  tests/agent/test_model_metadata.py \
  tests/tools/test_write_approval.py \
  tests/tools/test_file_write_safety.py \
  tests/tools/test_computer_use_delivery_ladder.py \
  tests/agent/transports/test_codex_app_server_session.py \
  -q
# 447 passed, 12 skipped
```

New coverage includes POSIX directory-sync ordering and failure propagation, blocked-hook and blocked-notifier AFK liveness, stale-request revocation, final grant revalidation, state permissions/symlink rejection, sidecar/multimodal cache safety, command routing, and Slack thread handling.

## Safety notes

- Disengaged by default; when no AFK state is active the turn-prologue path adds nothing and stays byte-stable for prompt caching.
- Engaging AFK never fabricates a user denial — the agent is told the absence is automatic.
- State and lock symlink leaves are rejected; writes use a no-follow temporary file plus atomic replacement and never modify a symlink victim. POSIX replace/unlink transitions fsync the containing directory and fail closed if that durability receipt cannot be produced. Native Windows makes no custom ACL or POSIX directory-fsync claim (documented in the slash-commands reference).
- Production-soaked: this has been running daily on a live fork (Slack gateway + CLI + desktop) since 2026-08-25, including through two upstream merge cycles.

### DIFF
--- a/agent/afk.py
+++ b/agent/afk.py
@@ -1,0 +1,687 @@
+"""Durable AFK (away-from-keyboard) availability state.
+
+``/afk`` — and Slack's thread-safe ``!afk`` — records that the operator has
+stepped away, so agent turns that happen while they are gone know not to wait
+on them and, just as importantly, know that nobody is there to approve
+anything.
+
+**One record for the whole machine.** Availability is a fact about the human,
+not about a Hermes instance, so the state lives at the Hermes *root* —
+:func:`hermes_constants.get_default_hermes_root`, resolved at call time — not
+under the active profile home. A ``/afk`` typed in the work profile is seen and
+clearable from the personal profile, the CLI, cron, and every gateway session.
+This is a deliberate exception to the profiles-are-islands rule: isolating
+availability per profile would mean the same person is simultaneously present
+and absent, which is exactly the confusion the command exists to remove.
+
+**Durability before success, serialized.** :func:`engage` holds a no-follow
+cross-process advisory file lock (``fcntl`` on POSIX, ``msvcrt`` on Windows)
+across write **and** readback, so
+the payload it returns is its own committed write rather than a neighbour's.
+The write itself is atomic (temp file + fsync + rename), so a concurrent reader
+never sees a torn document.
+
+**Permissions, stated truthfully.** On POSIX the file is written 0o600 and the
+mode is *verified afterwards*; a write that cannot be proven owner-only fails
+and removes the record. Native Windows has no POSIX mode bits and we install no
+DACL, so we claim nothing beyond the per-user ``%LOCALAPPDATA%`` location —
+see :func:`enforces_owner_only_permissions`.
+
+**Corrupt or partial state fails engaged.** The presence of the file is the
+status; its body is only metadata. A truncated, empty or hand-``touch``-ed file
+still reports AFK with unknown metadata, because the operator saying "I'm away"
+should not be undone by a bad byte.
+
+**The reason is untrusted data and never reaches the model.** It is bounded and
+neutralized on every *read* (not just on write — the file is bytes on disk that
+anything could have produced), and it is used only for the operator's own
+``/afk status`` reply. :func:`turn_context_note` carries status and timestamp
+only.
+
+**AFK is never authorization.** The note says so explicitly, and the shared
+approval boundary backs it with behaviour: while AFK, an approval prompt is
+never sent and every pending request is denied outright
+(:data:`APPROVAL_DENY_REASON`).
+
+**No identity is stored.** The state records availability and nothing else — no
+platform, no chat, no user id — so it cannot become a backdoor way to carry a
+Slack/Telegram identity into another surface. "The operator" means the owner of
+this machine's Hermes; gateway mutation requires an explicitly configured
+administrator, while a local CLI invocation already runs as the machine user.
+"""
+
+from __future__ import annotations
+
+import errno
+import json
+import os
+import stat
+import tempfile
+import threading
+from contextlib import contextmanager
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Any, Optional
+
+STATE_NAME = "afk.json"
+LOCK_NAME = "afk.lock"
+
+# Owner-only on POSIX. Availability plus a free-text reason is
+# personal-schedule data, and the Hermes root may be group-readable.
+STATE_MODE = 0o600
+
+# Upper bound on the reason. Chat-supplied text, bounded on read and write.
+MAX_REASON_CHARS = 200
+
+# Upper bound on the timestamp field read back off disk. Nothing legitimate
+# comes close; the cap exists because the field is read from a file.
+MAX_TIMESTAMP_CHARS = 64
+
+# The whole point of the note: an absent operator has not delegated authority.
+NO_AUTHORITY_SENTENCE = (
+    "AFK was a status, not a permission: at receipt time it did NOT widen "
+    "approvals, grant authority, or authorize any consequential or irreversible "
+    "action on the operator's behalf. Anything requiring approval remained "
+    "unauthorized unless it was separately approved."
+)
+
+# Deterministic denial handed to the tool that asked for approval while the
+# operator was away. Constant, not a rendering: the agent must get the same
+# explanation every time, and it must be unmistakable that being away is not
+# a decision about the request.
+APPROVAL_DENY_REASON = (
+    "Denied automatically: the operator is AFK and cannot answer an approval "
+    "request. AFK is availability, not approval — it is not consent and not a "
+    "judgement about this command. Do the safe, reversible parts of the task "
+    "and re-request this step once they are back."
+)
+
+APPROVAL_STATUS_UNKNOWN_REASON = (
+    "Denied automatically: Hermes could not safely verify the machine-global "
+    "AFK status, so it cannot establish that the operator is available to "
+    "approve this request. No consent was granted."
+)
+
+
+class AfkStateError(RuntimeError):
+    """The AFK state could not be durably written, verified, or removed."""
+
+
+def _afk_root() -> Path:
+    """Resolve the machine-wide Hermes root at call time.
+
+    ``get_default_hermes_root()`` maps ``<root>/profiles/<name>`` back to
+    ``<root>``, so every profile — and the default home — resolves to the same
+    directory. Freshness-correct: it re-derives from ``HERMES_HOME`` on each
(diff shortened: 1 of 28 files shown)


andrexibiza (commented):
Reviewed exact head `1b1bdb046a0e8a0d25237c6492283ff4df6c83bb` against exact base/current `main@4209d371aa1bb8840ce8447555bdd863a1a96c38` (1 ahead / 0 behind). I walked the state owner, the approval presentation/finalization paths, pending-queue settlement, the API-content sidecar, gateway command authority, tests, current Actions state, and the adjacent AFK/approval graph. There is a lot of careful work here: in particular, the final revalidation before persisting once/session/always grants, the explicit admin-only gateway mutation rule, corrupt-state fail-closed semantics, and keeping the operator's free-text reason out of model context are all the right direction.

I do see two P1 ownership/durability blockers before this should become the machine-global approval authority, plus one graph reconciliation that needs to be explicit.

### P1 — the claimed durable AFK bit is not crash-durable at the directory entry

`agent/afk.py::_atomic_replace_json()` fsyncs the temporary file and then calls `os.replace(temp_name, path)`, but it never fsyncs the containing directory. `clear()` similarly treats `path.unlink()` returning as durable proof and does not fsync the parent directory. That proves process-visible atomicity, not persistence of the rename/unlink across a host crash/power loss.

That distinction matters here because presence of `afk.json` is not ordinary metadata: it is the machine-global fact that gates whether interactive authorization may be presented/accepted. On a first `engage()`, the file contents can be fully fsynced while the directory entry that names them is still not durable; after a crash/restart Hermes can therefore come back with no AFK record and resume permitting approvals even though the last successful user-visible transition was “away.” The reverse direction can also resurrect a cleared record. The new state tests exercise readback, atomic visibility, concurrent threads, permissions, and symlink behavior, but none witnesses directory durability.

Required repair: on POSIX, fsync the parent directory after the successful `os.replace()` and after `unlink()` (and make failure part of the transition result rather than reporting success). Keep Windows semantics stated separately rather than pretending the POSIX recipe applies there. Add a regression that proves the directory-sync step is part of both commit and clear; ideally keep the state writer small enough that the ordering is mechanically obvious. The docstrings/tests should distinguish “atomic/read-back committed” from “crash-durable” unless that stronger receipt is actually produced.

### P1 — an observer/plugin can currently hold the global AFK transition lock hostage

The new `_await_gateway_decision()` takes `afk.status_transaction()` and then, while the cross-process `afk.lock` is still held, calls both `_fire_approval_hook("pre_approval_request", ...)` and `notify_cb(...)` before releasing the transaction.

The existing hook architecture makes this a real boundary problem, not a theoretical style concern. `_fire_approval_hook()` synchronously calls `hermes_cli.lifecycle.invoke_hook()`, which synchronously delegates into the plugin manager. The plugin manager's own timeout policy explicitly leaves `pre_approval_request` / `post_approval_response` outside the bounded-hook set because they were observers and “the approval UX already has its own timeout.” In other words, this PR moves code that was intentionally allowed to run without a hook timeout *inside the machine-global availability lock*.

A slow or hung approval observer can now prevent another thread/process from committing `/afk on`; on POSIX the other process blocks on `flock`, and on Windows the `msvcrt` locking path can fail its acquisition. That inverts the intended authority: plugin/notification liveness becomes a prerequisite for the operator's away-state safety transition. It also undercuts the PR's central mid-run use case — “step away while the agent is working” — exactly when an approval presentation hook may be active.

Required repair: keep arbitrary observer/transport work outside the machine-global status lock. The grant sink still needs exact final revalidation under the lock (the PR already does that well). For presentation, use a bounded admission/generation token or entry state that is created under the lock, release before observer/transport effects, and then re-check/revoke before any grant can become consumable. The important invariant is `AFK commit does not wait on plugin/network/UI liveness`, while a late presentation response can never become authority. Please add a torture test with a deliberately blocked `pre_approval_request` callback and a concurrent `engage()` proving AFK commits and drains/invalidates the pending approval without waiting for the observer to return.

### Graph / merge-order: this is not the only open owner of AFK state

The relation section correctly identifies #23804 as semantically complementary (idle follow-up turns vs explicit operator availability), but “no shared code paths” is not literally true: both PRs modify `gateway/run.py`, so they need ordinary rebase/composition ordering.

More importantly, open #21383 by `tymrtn` / issue #21122 predates this PR and already owns a machine-shared user-status record with an explicit `afk_status` field plus cross-profile per-turn injection. That work is broader and weaker for authorization (30-minute staleness and no approval gate), so I would **not** call it a duplicate of this PR. But two independently writable AFK truths cannot both be canonical. If both landed as written, `agent/user_status.py` and `agent/afk.py` could disagree about the same human's availability and inject contradictory context. Please reconcile that lineage explicitly: either this PR supersedes only #21383's `afk_status` slice while preserving `tymrtn`'s credit for the earlier cross-bot status design, or #21383 composes on the new AFK owner and drops its second availability store. Its other status fields remain separate/complementary work.

There are also direct file collisions with draft #98011 (`tools/approval.py`, `gateway/run.py`), whose Telegram master-task notification work is complementary rather than superseded. Whichever lands second needs to rebase and re-prove the approval payload/notification path; neither should silently overwrite the other's ownership changes.

### Exact-object acceptance

The submitted head is cleanly based on current main, which is good. Hosted acceptance is not green yet: exact-head CI `33276932277`, Docker `33276931907`, and Nix `33276931997` are all `action_required`; CI currently has zero jobs, so the locally reported `436 passed, 3 skipped` is useful construction evidence but not a hosted exact-head landing receipt.

This is substantial, thoughtful safety work. The final grant revalidation and the explicit “absence is never consent” contract are especially worth keeping. Fixing the crash-durability boundary and removing observer liveness from the global transition lock would make the mechanism match the strength of the claims around it. 🚀

lorenzespinosa:
Addressed review 5059279332 in exact head `7a00d63e09077b536c262d70e750920302ec30f3`.

- Added POSIX parent-directory fsync after AFK replace/unlink, including cleanup, with fail-closed error propagation and separate Windows semantics.
- Moved unbounded pre-approval hooks and notifier callbacks outside the machine-global AFK transaction while preserving locked admission, request-ID revocation, queue drain, and final grant revalidation.
- Added blocked-hook/notifier torture coverage proving AFK commits without waiting and late responses cannot authorize.
- Reconciled canonical AFK ownership with #21383 and merge ordering with #23804/#98011 in the PR body.

Verification: focused `82 passed, 12 skipped`; full affected suite `447 passed, 12 skipped`; Ruff/py_compile/diff-check clean. Independent Kimi exact-SHA review: C0/H0/M0/L2 (both Low findings are fail-closed status-message mismatches only).

lorenzespinosa:
Superseded by focused PR #101233, which isolates durable machine-global AFK state and shared command wiring from approval and model-context changes. The broader branch remains preserved for follow-up work.

PR #81554, opened 2026-08-08
feat: embeddable durable-execution layer + mandatory egress guardrails

dhyabi2:
# feat: embeddable durable-execution layer (journal/outbox) + mandatory egress guardrails

Branch: `feat/durable-execution-runtime` (4 commits, each one logical change)

## Problem

Hermes has strong pieces of durability, but they don't compose into durable-execution guarantees:

1. **Outbound delivery is durable on only one of four send paths.** `gateway/delivery_ledger.py` (idempotency keys, owner-liveness claims, attempt caps) is wired only into the gateway final-response path. Cron/`DeliveryRouter`, the `send_message` tool, and outbound webhooks have zero delivery durability.
2. **No egress guardrail exists.** `agent/redact.py` runs on ingress/display/snapshots — never on the bytes sent to Telegram/Slack/Discord. A secret in a reply is delivered verbatim on every path: streaming edits, captions, cron, `send_message`, ledger redelivery.
3. **Mid-tool crashes degrade to "effect UNKNOWN".** `agent/replay_cleanup.py` infers state from a torn transcript instead of replaying recorded intent; there is no journal.
4. **`synchronous=FULL` was macOS-only** — on Linux, committed transcript turns can be lost on power loss.
5. **No crash-injection tests** anywhere in the tree.

## The fix, in two diagrams

**Where the egress boundary sits.** Every in-process send path bottoms out in the wrapped adapter methods (`BasePlatformAdapter.__init_subclass__` wraps `send`/`edit_message` on all 22 platform adapters + Relay), so no caller can bypass it. The two paths that never reach a wrapped method get their own guard:

```mermaid
flowchart TD
    SC["stream consumer<br/>(8 direct send/edit sites)"] --> W
    SWR["_send_with_retry<br/>(final replies + plain-text fallback)"] --> W
    CAP["captions / clarify / voice /<br/>private notices (base.py self.send)"] --> W
    LED["delivery-ledger boot redelivery<br/>(gateway/run.py adapter.send)"] --> W
    DR["cron / DeliveryRouter"] -->|"native adapter"| W
    DR -->|"relay: send_for_platform"| G2["guard_outbound_text<br/>(full guard)"]
    SMT["send_message tool"] -->|"guard whole body pre-chunk,<br/>then live adapter"| W
    SMT -->|"standalone raw senders<br/>(telegram/signal/matrix/...)"| G3["pre-chunk redaction<br/>(only boundary for this path)"]
[...]

PR #72144, opened 2026-07-26
fix(gateway): dedup provider-error status and final response for adapters without send_or_update_status

x7peeps:
## 背景

修复 #72131: Provider errors delivered twice on adapters without send_or_update_status (mid-run status rewrite + identical final response)

## 根因分析

在平台适配器未实现可选方法 `send_or_update_status` 时，`_send_or_update_status_coro` 会降级为调用 `adapter.send()` 发送状态消息。当 provider 发生错误时，两条通道都会发送相同的改写后文本：

1. 运行中状态回调 → `_prepare_gateway_status_message` 改写后通过 `_send_or_update_status_coro` 发送（降级为 `send()`，成为持久消息）
2. 失败的 turn 最终响应 → `_sanitize_gateway_final_response` 改写后再次发送（相同的用户友好文本）

对于 Telegram 等实现了 `send_or_update_status` 的适配器，状态气泡会被原地编辑并在运行结束后清理，所以只有最终消息保留。但对于没有该方法的适配器，状态消息变成第二条持久消息，导致用户看到两条相同的错误提示。

## 修复方式

在 `gateway/run.py` 中新增 dedup 机制：

- 新增模块级变量 `_last_sent_provider_error_status` 跟踪通过 plain-send 降级路径发送的 provider 错误状态文本
- 在 `_send_or_update_status_coro` 降级路径中，如果内容是 provider 错误回复，记录该文本
- 在 `_sanitize_gateway_final_response` 中，如果改写结果与已发送的状态文本相同，返回空字符串跳过重复发送
- 在调用方（final_response 为空时的回退逻辑）中，检查 `_deduped_provider_error` 标志，避免回退到原始错误文本

## 验证

- [x] 代码变更已在本地验证
- [x] 遵循项目 AGENTS.md 贡献规范
- [x] 无破坏性变更
- [x] 新增 21 个单元测试全部通过
- [x] 现有 617 个 noise filter 测试全部通过
- [x] 现有 144 个 status 相关测试全部通过

Closes #72131

PR #47791, opened 2026-06-17
fix(gateway): avoid duplicate threaded provider errors

potatogim:
## What does this PR do?

Provider/API failures can reach the user twice on Telegram: once via the threaded status callback, again via the final sanitized response. This PR suppresses the duplicate by tracking the status delivery future — the final send is skipped only when the threaded status was delivered. If the status send fails or times out, the final response still goes through.

Also fixes `send_or_update_status()` so its cache key includes `thread_id` and `dm_topic_id`, preventing status bubbles from one topic editing a bubble in another. Adds per-key `asyncio.Lock` to serialize concurrent calls with the same routing key.

## Related Issue

N/A — found while investigating duplicate Telegram provider error messages.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- **`gateway/turn_context.py`**
  - Add `_provider_error_status_futures: List[tuple]` field to `TurnContext`.
- **`gateway/run.py`**
  - `TurnRunner._status_callback_sync`: detect provider errors, append matching send futures to `ctx._provider_error_status_futures`.
  - `_provider_error_status_already_delivered()` — awaits matching future (2s timeout), returns True on confirmed delivery.
  - Final-response suppression: skip only if threaded status delivery succeeded; on failure/timeout, normal delivery path runs.
  - `already_sent` guard: allow `failed + status_provider_error_delivered` (ordinary failures always delivered).
- **`plugins/platforms/telegram/adapter.py`**
  - `send_or_update_status()` cache key: `(chat_id, status_key)` → `(chat_id, status_key, thread_id, dm_topic_id)`.
  - `_status_locks: Dict[tuple, asyncio.Lock]` — per-key lock for concurrent-call serialization.
- **`tests/gateway/test_telegram_status_update.py`**
  - Thread isolation, DM-topic isolation, edit-in-place within same topic.
  - `test_concurrent_same_key_is_serialized` — event-gated async verifying two concurrent calls serialize.
- **`tests/gateway/test_gateway_silence_tokens.py`**
  - Ordinary failed responses delivered even when `already_sent=True`.
  - `status_provider_error_delivered=True` suppresses duplicate final send.

## How to Test

1. Run targeted tests:

   ```shell
   python -m pytest tests/gateway/test_telegram_status_update.py tests/gateway/test_gateway_silence_tokens.py -q
   ```

   Result: `18 passed`

2. Compile changed files:

   ```shell
[...]

PR #98165, opened 2026-08-29
fix(gateway): suppress the duplicated provider-error status copy

### LINKED ISSUE #72131: Provider errors delivered twice on adapters without send_or_update_status (mid-run status rewrite + identical final response)
### Summary
On platform adapters that do not implement the optional `send_or_update_status`, a provider failure reaches the chat **twice** as two identical persistent messages — e.g. the user sees
`⏱️ The model provider is rate-limiting requests. Please wait a moment and try again.` ×2 for one failed turn.

### Why it happens
A provider failure is legitimately surfaced through two gateway channels:
1. the mid-run status callback → `_prepare_gateway_status_message` rewrites the raw provider error into the user-safe text and sends it;
2. the failed turn's final response → `_sanitize_gateway_final_response` produces the **same** rewritten text and sends it again.

Adapters that implement `send_or_update_status` (e.g. Telegram, #30045) are only *partially* protected: the in-place edit collapses the mid-run rewrites into a single status bubble, but that bubble is only removed by the temp-bubble cleanup path, which is opt-in per platform (`display.platforms.<platform>.cleanup_progress`) **and** explicitly skipped on failed runs (`and not response.get("failed")` — "Failed runs skip cleanup so the bubbles remain as breadcrumbs"). A provider failure *is* a failed turn, so the polished status bubble survives alongside the byte-identical final response — see the production repro in [this comment](https://github.com/NousResearch/hermes-agent/issues/72131#issuecomment-5465060605). *(This section originally claimed Telegram was unaffected; corrected per that repro.)*

For any adapter without the optional method it is worse still: `_send_or_update_status_coro` falls back to plain `send()` — the transient status becomes a second persistent message with different message ids, so downstream idempotency can't dedup it.

### Suggestions (either would do)
- Gateway-side: when a status message was rewritten by `_gateway_provider_error_reply` and the turn then ends with an identical rewritten final response, skip one of the two (dedup identical provider-error text per chat within a turn); or
- Docs/contract: state explicitly that chat-surface adapters should implement `send_or_update_status` (and what semantics the gateway expects — transient, superseded, cleaned up), since without it every mid-run status degrades to a persistent message. Note this only shrinks the blast radius: per the failed-turn cleanup gate above, even a conforming adapter still shows the duplicate on provider failures, so the gateway-side dedup is the fix that covers both cases.

### Workaround we ship
Our platform adapter now implements `send_or_update_status` and routes all statuses to the platform's ephemeral (replaceable, never persisted) bubble — that fully fixes the duplication on our side. Filing this because every third-party adapter without the optional method hits the same footgun (and, per the above, even adapters with it hit the failed-turn variant).

liuhao1024:
## What does this PR do?

Fixes the double delivery of provider errors to chat surfaces (#72131). A provider failure was surfaced through two gateway channels that produce byte-identical text:

1. the mid-run status callback — `_prepare_gateway_status_message` rewrote the raw provider error via `_gateway_provider_error_reply` and sent it;
2. the failed turn's final response — `_sanitize_gateway_final_response` rewrote it to the same user-safe text and sent it again.

For adapters without `send_or_update_status`, `_send_or_update_status_coro` falls back to a plain `send()`, so the transient status becomes a second persistent message. And per ValeryP's analysis on the issue, even Telegram is affected on failed turns: progress-bubble cleanup is opt-in per platform (`display.platforms.<platform>.cleanup_progress`) and its registration is guarded by `and not response.get("failed")` ("Failed runs skip cleanup so the bubbles remain as breadcrumbs") — a provider failure is a failed turn, so the surviving bubble is a verbatim copy of the final answer.

The fix suppresses the transient status copy instead of delivering the rewrite: a provider-error-shaped status returns `None` from `_prepare_gateway_status_message`. This closes both surfaces with a single stateless change — no turn-scoped dedup state (the scoping problem that sank #72144). The final response still reports the failure with the same sanitized text, so the user still sees exactly one user-safe message per failed turn, and no raw provider body is delivered on either path.

## Related Issue

Fixes #72131

## Type of Change

## Changes Made

- `gateway/run.py` — `_prepare_gateway_status_message`: provider-error-shaped text now returns `None` (suppressed) instead of being rewritten and delivered; the rewrite previously duplicated the final response's text as a persistent message.
- `tests/gateway/test_telegram_noise_filter.py` — replaced `test_telegram_status_sanitizes_raw_provider_security_errors` (asserted the rewrite was delivered on the status path) with `test_chat_gateways_suppress_provider_error_status` parametrized over chat platforms, asserting suppression; added `test_local_status_keeps_raw_provider_errors` pinning that local/CLI raw-text surfaces keep the raw diagnostic stream.

## How to Test

1. `python3 -m pytest tests/gateway/test_telegram_noise_filter.py tests/gateway/test_local_model_connection_reply.py tests/gateway/test_compression_progress_notices.py -q` — Observed result: 183 passed.
2. `python3 -m pytest tests/gateway/ tests/agent/test_turn_context_overflow_warning.py -q` — Observed result: only the pre-existing `tests/gateway/relay/test_relay_going_idle.py` timing flakes fail; they fail identically on a clean `upstream/main` checkout (4 failed there), unrelated to this diff.
3. Behavioral check of the duplication itself: `python3 -c "from gateway.run import _prepare_gateway_status_message, _sanitize_gateway_final_response; from gateway.config import Platform; raw='API call failed after 3 retries: HTTP 429'; print(_prepare_gateway_status_message(Platform.TELEGRAM, 'lifecycle', raw)); print(_sanitize_gateway_final_response(Platform.TELEGRAM, raw))"` — Observed result: `None` (no status copy) followed by the single sanitized final reply, i.e. exactly one message per failed turn instead of two.

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate (#72144 is closed; #81554 is an unrelated durability/egress feature)
- [x] I've run `pytest tests/ -q` and all tests pass (gateway + related suites; only pre-existing main flakes fail, see How to Test)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (inline comment documents the suppression contract)

## For New Skills

N/A — not a skill.

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -878,7 +878,14 @@ def _prepare_gateway_status_message(platform: Any, event_type: str, message: str
         ):
             return None
     if _looks_like_gateway_provider_error(text):
-        return _gateway_provider_error_reply(text)
+        # A provider error also reaches the chat as the failed turn's final
+        # response, which _sanitize_gateway_final_response rewrites to the
+        # same user-safe text — delivering the rewrite here too means two
+        # identical persistent messages (#72131). Failed runs are exempted
+        # from progress-bubble cleanup, so even adapters that implement
+        # send_or_update_status keep the status copy. Suppress the transient
+        # copy; the final response still reports the failure.
+        return None
     return text
 
 
--- a/tests/gateway/test_telegram_noise_filter.py
+++ b/tests/gateway/test_telegram_noise_filter.py
@@ -255,20 +255,29 @@ def test_chat_gateways_drop_interrupt_sentinel(platform):
     assert _sanitize_gateway_final_response("local", sentinel) == sentinel
 
 
-def test_telegram_status_sanitizes_raw_provider_security_errors():
-    """Provider policy/security bodies should be replaced before chat delivery."""
+@pytest.mark.parametrize("platform", CHAT_PLATFORMS)
+def test_chat_gateways_suppress_provider_error_status(platform):
+    """Provider-error statuses must not double up with the final response.
+
+    The failed turn's final response is rewritten to the same user-safe text
+    by `_sanitize_gateway_final_response`, so a status copy here would land as
+    a second identical persistent message (#72131). Suppression also keeps the
+    raw body (request ids, policy text) out of chat — nothing is delivered on
+    this path at all.
+    """
     raw = (
         "❌ API failed after 3 retries — HTTP 400: request blocked because "
         "Operation contains cybersecurity risk. request_id=req_123"
     )
 
-    sanitized = _prepare_gateway_status_message(Platform.TELEGRAM, "lifecycle", raw)
+    assert _prepare_gateway_status_message(platform, "lifecycle", raw) is None
 
-    assert sanitized is not None
-    assert "provider rejected" in sanitized.lower()
-    assert "cybersecurity risk" not in sanitized.lower()
-    assert "HTTP 400" not in sanitized
-    assert "req_123" not in sanitized
+
+def test_local_status_keeps_raw_provider_errors():
+    """Local/CLI surfaces keep the raw diagnostic stream on the status path."""
+    raw = "API call failed after 3 retries: HTTP 429 Too Many Requests"
+
+    assert _prepare_gateway_status_message("local", "warn", raw) == raw
 
 
 def test_telegram_final_response_sanitizes_raw_provider_errors():


alt-glitch:

Related to #47791: both address duplicate provider-error delivery for #72131, but this PR suppresses the transient status copy while #47791 preserves a delivered status and suppresses an equivalent final response.

liuhao1024:
The Windows-only failure in this run is a known flaky timing test, unrelated to this diff.

**Failing test:** `tests/test_desktop_update_windows_progress.py::test_progress_advances_while_the_orchestrator_blocks` — `AssertionError: /progress unresponsive until deadline (last error: TimeoutError('timed out'))`.

**Why it's unrelated to this change:**
- This PR only touches `gateway/run.py` (suppressing the duplicated provider-error status copy) and `tests/gateway/test_telegram_noise_filter.py` — no shared code path with the Desktop update `/progress` socket/HTTP chain.
- The same test failed with the identical `/progress unresponsive until deadline` signature on an unrelated PR the day before (#97462, kanban-only diff, run 33212153925), and this test has a documented deflake history on main (6d1284a073 Aug 19 "deflake the Windows progress self-test", 18a15a46d8 Aug 20 "survives transient /progress socket stalls") — the transient-stall race can still slip through under runner load.
- Everything else in the same run is green: `Python tests / Run tests`, `Python tests / e2e`, macOS-only tests, and all lints passed; only this deadline-sensitive test timed out.

No rerun attempted since external contributors can't trigger one; happy to rebase if a fresh SHA is preferred.

Issue #95798, opened 2026-08-26
fix(desktop): chat edit-message silent-fail races + model-switch jank


dlgrv:
## Summary

The chat edit flow (edit a sent user message, then resubmit) has known silent-fail races that are hard to reproduce by hand. This issue tracks fixing them, using the debug MCP server from #95489 as a mechanical reproduction harness.

## Symptoms (bug class)

1. **Blur-cancel kills the composer core before submit.** Opening the edit composer and losing focus (e.g. a stray click, window blur, or an incoming event) tears down the edit state without saving. The user's edit is silently dropped.
2. **Empty catch swallows "Composer is not available".** `apps/desktop/src/app/chat/composer/user-edit-composer.tsx` (around the submit path) catches and discards the error instead of surfacing it, so a failed resubmit looks like "nothing happened" with no console error and no rollback explanation.
3. **Model-switch row jank.** Switching the model in an existing thread causes layout shift / flicker that is currently unquantified and unguarded.

## Why fix separately from #95489

#95489 is the *tooling* (an MCP server that lets an agent drive the UI). This issue is the *bugfix* in the chat code itself. Keeping them separate means the tooling can land (or be reviewed) independently of the product fix, and the fix PR stays focused — one logical change per PR, per CONTRIBUTING.md.

## Reproduction harness

The `ui_flow_edit` tool built in #95489 reproduces this mechanically against a live renderer:

- opens the edit composer on the last user message
- types a replacement
- presses Enter
- reports `sendAccepted` / `composerCoreAlive` / `timelineChanged` + console errors

This turns "flaky UI bug" into a repeatable probe and a future regression test.

## Proposed fix direction

- Replace the empty catch in the edit-submit path with a visible failure (toast / rollback with explanation) instead of silent drop.
- Guard the edit composer against blur-teardown racing the submit (commit/flush before unmount, or disable submit-on-blur).
- Quantify model-switch layout shift and debounce the row swap.

## Test plan

- [ ] `ui_flow_edit` repro produces a clear failure signal (not silent drop) after the fix.
- [ ] Manual: edit a sent message, click away mid-edit, return — edit is not lost without explanation.
- [ ] Model-switch in a populated thread shows no visible layout jump.

## Related

- Tooling: #95489 (Debug MCP server, includes `ui_flow_edit`)
- PR with the harness: #95781

SANTHOSHKUMAR-303:
I understand the issue clearly .Could you please assign it to me so I can work on it?

[...]

PR #98166, opened 2026-08-29
test(desktop): pin dirty edit blur→Enter interleaving (#95798)

dlgrv:
## Summary
- Pins the `#95798` dirty-edit blur→Enter interleaving that was previously untested (dirty draft → blur arms the 80ms cancel timer → Enter must still submit; and Enter after a survived blur).
- On current `main` both cases are already green — the silent-fail class was closed by earlier work (dirty-blur guard, timer drain on unmount, macOS send-arrow gesture test). This PR only adds the missing regression coverage so the hole cannot reopen quietly.
- Leaves the **model-switch jank** half of #95798 untouched (separate deliverable).

## Test plan
- [x] `cd apps/desktop && npx vitest run src/components/assistant-ui/thread/user-message-edit.test.tsx` → **12/12**
- [x] Existing `user-message-edit-gesture.test.tsx` still green

Partial progress on #95798 (edit-race half); model-switch jank remains open.

Made with [Cursor](https://cursor.com)


Enough1122:

**Overall:** Pure regression coverage for #95798 blur→Enter interleaving. No prod code change; strengthens guardrails around the previously fixed silent-fail / latch bug.

**What it does**
- `apps/desktop/src/components/assistant-ui/thread/user-message-edit.test.tsx:14` adds two tests inside `Enter submission and latch behavior`: (1) dirty edit → `blur` arms 80ms cancel timer → immediate `Enter` still submits (waits 200ms, asserts `onEdit` once, no `Composer is not available` uncaught), (2) dirty edit survives blur (`moveFocusOutside` leaves composer mounted) then `Enter` submits.
- Documents the untested hole ("dirty-blur guard, timer drain on unmount, macOS send-arrow gesture" already green) and pins it.

**Non-blocking notes**
- First test hooks `process.on('uncaughtException')` stack — ensure `process.off` in `finally` runs even if `waitFor` times out, to avoid leaking handler across suite.
- Uses `window.setTimeout(200)` + `waitFor` — race-sensitive but matches the 80ms timer window; no assertion on timer drain itself (covered elsewhere), so this is strictly a pin.

No blocking issues.

PR #98167, opened 2026-08-29
fix(update): stop reporting a carried local commit as "update available"

andredezzy:
## What

A shallow clone with a carried local commit reports "Update available" forever. Running `hermes update` cannot clear it. This makes both update checks ask git before the GitHub compare API.

## Reproduce

On a shallow clone (what the installer produces), with any commit that exists only locally — a local patch, a cherry-picked fix, a branch carried across updates:

```
0
0                                    # upstream tip is an ancestor: we are AHEAD

$ hermes update --check
→ Fetching from origin...
```

`~/.hermes/.update_check` disagrees with the command in the same session:

```json
{"ts": 1788040488.4, "behind": 0, "rev": null, "ver": "0.20.6"}
```

## Cause

A shallow clone cannot count across its boundary, so both checks compare tip SHAs and, when they differ, ask the compare API for the distance. A local-only commit is not on GitHub, so the API 404s:

```python
>>> _github_compare_behind(head_sha, upstream_sha)
None
```

Both call sites read `None` as "behind, count unknown":

- `hermes_cli/banner.py` `_check_via_local_git` shallow branch → `UPDATE_AVAILABLE_NO_COUNT`
- `hermes_cli/update_cmd.py` `--check` shallow branch → `"⚕ Update available (behind origin/main)"`

Updating cannot fix it. The carried commit is still there afterwards, still invisible to the API, so the notice comes back every time — and the only action it suggests is the one that risks the carried work. `banner.py` already warns about exactly this failure mode on the non-shallow SSH path:

> Tip SHAs alone can't distinguish "behind" from a local carried commit sitting AHEAD of origin/main, and misreporting an ahead checkout as behind nudges the user into `hermes update`, which can wipe their carried work.

That reasoning holds on the shallow path too; the check was just never added there.

## Fix

Ask git before the API:

```python
ancestor = subprocess.run(
    ["git", "merge-base", "--is-ancestor", target_rev, "HEAD"], ...
)
if ancestor.returncode == 0:
    return 0
```

`merge-base --is-ancestor` resolves from the fetched objects even on a shallow clone, and git is the only party that knows about a commit which exists nowhere else. Applied to both sites, mirroring the SSH fast path that already does this.

A genuinely behind checkout is unaffected: the tip is not an ancestor, so it falls through to the compare API for the exact count, or to the honest no-count sentinel when the API is unreachable.

## Tests

Two added to `tests/hermes_cli/test_banner_git_state.py`, written RED first (the first failed with `assert -1 == 0`, the sentinel this PR removes):

- `test_shallow_local_ahead_is_not_reported_as_behind` — ancestor check passes, API returns `None`, result is `0`
- `test_shallow_genuinely_behind_still_reports_the_sentinel` — ancestor check fails, API returns `None`, sentinel preserved

## Verification

- `pytest tests/hermes_cli/test_banner_git_state.py test_banner.py test_update_check.py test_cmd_update.py` — 55 passed
- End to end on the checkout that surfaced this: `hermes update --check` goes from `⚕ Update available (behind origin/main)` to `✓ Already up to date.`

I have not run the full `tests/` suite end to end and am not claiming a green run I did not watch finish.


liyangbing:
<h3>Review</h3><p>This PR fixes the important directionality error in both the local banner path and <code>update --check</code>: differing tip SHAs are not sufficient evidence that the checkout is behind, and the local <code>merge-base --is-ancestor</code> check can recognize a carried commit that the compare API cannot see. The desktop SSH path also applies the same rule through <code>resolveSshBehindCount</code>.</p><h3>Suggested contract</h3><ul><li>Keep the three outcomes distinct: target is an ancestor of <code>HEAD</code> means up to date/ahead, a positive compare count means behind, and an inconclusive ancestry/API result means unknown—not update available.</li><li>Add direct regression coverage for the changed <code>hermes_cli/update_cmd.py</code> shallow <code>--check</code> branch. The current Python tests exercise <code>banner.py</code>, but the second production caller can regress independently.</li><li>Exercise the desktop <code>checkUpdates()</code> integration with a local-only commit, a genuinely behind checkout, an unavailable <code>merge-base</code>, and a failed <code>ls-remote</code>; verify that the UI state and persisted update marker agree.</li><li>Define timeout/error handling for <code>merge-base</code> consistently across Python and Electron. A non-zero result currently covers both “not ancestor” and execution failure, so the latter must not be silently treated as proof of being behind.</li><li>Test a fetched remote tip that is an ancestor of <code>HEAD</code> with multiple local commits and a shallow boundary, plus the inverse case where the target is not reachable and the compare API returns an exact count.</li></ul><h3>Regression matrix</h3><ul><li>plain equal tips, local-ahead shallow checkout with a local-only commit, and genuinely behind shallow checkout</li><li>Python banner path and Python <code>update --check</code> path, including their output and cached marker behavior</li><li>desktop SSH-official path with exact count, unknown count, local-ahead, and equal-tip cases</li><li>merge-base timeout/launch failure versus a valid non-ancestor result</li><li>failed remote lookup, changed branch target, and update UI state after a carried commit survives an attempted update</li></ul><p>The reported 55 focused passes and end-to-end before/after result are strong evidence. Covering the second Python caller and distinguishing Git execution failure from a valid ancestry-negative result would make the safety guarantee complete.</p>

andredezzy:
Both points were fair, and both are now addressed in `9fbaabf`.

**Direct coverage for the `update_cmd.py` branch.** You were right that `banner.py` and `update_cmd.py` are two independent callers of the same rule and only the first had tests. `tests/hermes_cli/test_update_check_ancestry.py` drives `_cmd_update_check()` against a stubbed shallow checkout and covers all three outcomes: carried local commit reads as up to date, genuinely behind still reports the count, and a failed probe does not report "update available".

**The conflated non-zero exit.** This was a real defect, not just a documentation gap. `merge-base --is-ancestor` returning non-zero means "not an ancestor"; git failing to execute means the question was never answered. The old code treated both as evidence and fell through to the compare API. Now an execution failure sets `ancestry_known = False`, and when the compare API also cannot answer, the command prints:

```
⚕ Update status unknown (could not compare with origin/main).
```

rather than claiming an update is available. That asymmetry is deliberate — misreporting an ahead checkout as behind nudges the user into an update that can discard the work they are carrying, so an unanswered question stays unanswered.

The probe also gained a 5s timeout, so a wedged git cannot stall `--check`.

Verified RED→GREEN: with the fix reverted, `test_a_failed_ancestry_probe_is_not_reported_as_behind` and `test_the_ancestry_probe_runs_under_a_timeout` both fail; with it, `tests/hermes_cli/test_update_check_ancestry.py` and `test_banner_git_state.py` are 11 passed.

andredezzy:
Went back through the review point by point — the Electron half of item 4 was still open, and it was hiding a worse bug than the contract mismatch. Pushed as `7a78c90`.

**`merge-base` timeout/error handling across Python and Electron.** The Python side landed in `9fbaabf`. On the Electron side I had only checked the contract and missed that `runGit` **rejects** on spawn failure (`child.once('error', reject)`) rather than exiting non-zero, and the ancestry probe had no catch. So a machine with a broken or missing git did not misreport the count — it threw out of `checkUpdates()` entirely. That is now three explicit states:

```ts
true   // proven ancestor — a carried local commit puts HEAD ahead
false  // proven not an ancestor
null   // git could not answer
```

Only the first two are evidence about the checkout, matching the contract you set out in item 1.

The rule moved into `resolveAncestry` in `update-count.ts` because `main.ts` is not importable at 17k lines; its caller now uses it instead of repeating the logic inline.

**On method, since it is relevant to how much these tests are worth.** My first attempt at covering this asserted against `resolveSshBehindCount` and passed *with the fix reverted* — `null` was already falsy there, so the assertion proved nothing. I threw it away and tested the behaviour instead. Removing the catch now fails `a git that cannot run yields unknown ancestry, not a verdict`.

Current state: `update-count.test.ts` 27 passed, `tests/hermes_cli/test_update_check_ancestry.py` + `test_banner_git_state.py` 11 passed, typecheck clean. One unrelated pre-existing failure in `managed-ssh-update.test.ts` (exit 127, reproduces on a clean tree).

**Still open from your matrix:** the full `checkUpdates()` integration test with the persisted update marker (item 3). That needs Electron IPC harnessing rather than a pure-function test, and I would rather say it is not done than imply the matrix is complete.

andredezzy:
I rechecked the Electron error path and found one caller still bypassing the three-state resolver. The fetched/HTTPS path called `runGit(...).code` directly, so a Git spawn failure could still escape `checkUpdates()` outside the official-SSH branch. Commit `b4bae080` now routes both paths through `resolveAncestry`; it also fixes the existing import-order lint failures in the touched files.

Verified: `update-count.test.ts` 27 passed, the two Python ancestry suites 11 passed, Electron typecheck passed, and ESLint passed for the touched files.

On the requested persisted-marker test: the desktop does not persist update-availability state. `$updateStatus` lives in memory, `updates.json` stores branch configuration, and `update-marker.ts` tracks updater handoff rather than availability. I did not add a test for persistence that the product does not implement. The renderer suite already covers `behind: null` with `updateAvailable: true`; a full `main.ts` IPC integration harness remains separate infrastructure work.

andredezzy:
Closed the remaining integration gap in `a59f69a8`. The new Playwright spec drives the real preload IPC and main-process `checkUpdates()` path against a temporary bare remote, a depth-one checkout, and a carried local commit. It proves that differing tips return `behind: 0` and `updateAvailable: false` when the remote tip is already an ancestor of `HEAD`.

The same commit tightens the unknown case: a failed ancestry probe plus an unavailable compare result no longer becomes an update prompt without evidence that the remote has commits the checkout lacks.

Verified: the E2E spec passed, `update-count.test.ts` passed 28 tests, Electron and E2E typechecks passed, ESLint passed for the touched files, and the desktop build completed.

andredezzy:
Merged current upstream and ported the ancestry checks onto the extracted update helpers.

The banner still collapsed Git execution failures into “update available.” New tests reproduced this for exit 128, timeout, and launch failure. Both Python callers now leave the result unknown when Git and the compare API cannot answer, while preserving exact counts and the existing no-count sentinel after a valid negative ancestry result.

Validation: 46 Python tests passed, including a real depth-1 checkout with two carried commits; 28 desktop update-count tests passed; Python lint, desktop ESLint, TypeScript, and the desktop build passed. The build emitted React Compiler/import-splitting warnings. Electron E2E was not rerun in this refresh.

andredezzy:
Closing this: current `main` already carries the fix.

`_check_via_local_git` in `hermes_cli/banner.py` now asks git before the compare API, and `_tips_behind` takes `repo_dir` so a fetched tip already reachable from HEAD reads as up to date. The desktop's `ls-remote` path performs the same check in `checkUpdatesViaLsRemote`. The premise this PR was opened on no longer holds, and the branch sits 2377 commits behind, so rebasing it would deliver nothing.

One gap remains, in code that landed after this PR. The API-first passive check added in "passive update checks use the GitHub API once a day, never git fetch" routes GitHub origins through `checkUpdatesViaApi`, which has no ancestry fallback. On a checkout carrying a local commit, HEAD exists nowhere on GitHub, the compare endpoint returns 404, `parseCompare` returns null, and the function reports `updateAvailable: true` with an unknown count. That is the same false positive on a new path, and its only suggested action is the one that can park the carried work.

Reproducing it needs only a local commit:

```
$ curl -s -o /dev/null -w '%{http_code}\n' \
    "https://api.github.com/repos/NousResearch/hermes-agent/compare/$(git rev-parse HEAD)...$(git rev-parse origin/main)"
404
0
```

I will open a separate PR for that path rather than repurpose this branch. Thanks to everyone who reviewed here.

Issue #79017, opened 2026-08-05
prompt_cache_key loses continuity across context-compression session rotation (needs a logical cache-scope concept)


JoaoMarcos44:
## Context

Follow-up from #78941 / PR #78959. Explicitly called out in that PR's description as an accepted trade-off, opening this issue so maintainers can decide whether/how to pursue a real fix, per the investigation report's recommendation.

## The gap

`_cache_scope_from_session_id(session_id)` (PR #78959) scopes `prompt_cache_key` by the **physical** session id. That's correct for the bug #78959 fixes (unrelated sessions sharing a bucket) and for cron (repeat fires of the same job normalize to one scope). But context-compression rotation mints a **new physical session_id mid-conversation** to segment the transcript. Since the cache scope is derived directly from that physical id, rotation always produces a new scope — the cache goes cold at every rotation boundary, even though it's logically the *same conversation* continuing.

```
root-session    -> pck_b2bfe5458decf53e5752b393
rotated-session -> pck_71b8ffefb95330d6cd90f3d3
same conversation, different keys: true
```

## Why this isn't a small patch

Three identities get conflated today and need to stay distinct:

| Identity | Correct role |
|---|---|
| Physical session/transcript id | Identifies one execution/segment; changes on `/new`, branch, and compression rotation |
| Logical cache-scope (conversation identity) | Should stay stable *within* the same conversation — inherited across compression rotation, but still reset on `/new`, branch, and per independent subagent |
| Provider sticky-routing affinity (OpenRouter `session_id`, xAI `x-grok-conv-id`) | Pins requests to the backend holding the warm cache; contract differs per provider |

A real fix needs a canonical "cache scope" helper that:
- inherits scope across compression rotation (walk to the pre-rotation session/parent),
- still creates a fresh scope on `/new`, on branch, and per independent subagent,
- doesn't reuse `gateway_session_key` or a generic conversation-root walk as a substitute (both outlive `/new` or cross branch boundaries incorrectly).

That's a genuine design task (finding/threading a stable "conversation identity" through the compression/rotation/branch code paths), not a one-line change like #79012–#79015.

## Ask

Filing for maintainer input on whether this is worth the redesign, and if so, whether the "logical cache-scope helper" sketched above (and in the referenced investigation) is the right shape.

JoaoMarcos44:
Demonstration PR recreated standalone (previous one dragged in #78959's unmerged diff, making it unreviewable): #79036

[...]

PR #98170, opened 2026-08-29
fix(cache): honor host-declared session key and conversation epoch for prompt cache affinity

StanleyStetson:
## Summary

Unifies and finalizes host-declared conversation affinity caching across all supported transports (OpenAI, Codex, OpenRouter, Nous, Grok), resolving the per-response session churn described in #96811 while building on the foundational work from #97158 and #97709:

1. **Host-Declared Affinity Scope**: Consumes \gateway_session_key\ (\X-Hermes-Session-Key\ / \uild_session_key\) and mixes \conversation_epoch\ to compute deterministic, collision-safe \gwk_<sha256(key:epoch)[:24]>\ affinity hashes.
2. **Monotonic Epoch Advancement**:
   - Explicit \/new\ (\SessionStore.reset_session()\) increments \conversation_epoch = old_epoch + 1\.
   - Policy auto-resets (\idle\, \daily\, and stale recovery in \get_or_create_session()\) monotonically advance \conversation_epoch = old_epoch + 1\, strictly preventing ABA roll-backs to epoch 1.
3. **SessionStore Abstraction Seam**: Added public \SessionStore.get_conversation_epoch(session_key: str) -> int\ helper consumed cleanly by \gateway/run.py\ without breaching store lock boundaries.
4. **Explicit Fork & Background Exclusions**: Explicit fork children (\/branch\, delegates, tool-spawned sessions) and background review forks (\_persist_disabled\) isolate to their own physical session scopes per #79161 invariants.
5. **ContextVar Propagation & Nested Turn Shadowing**: Manages \_affinity_scope\ via task-local ContextVar with safe \	ry/finally\ lifecycle handling in \
un_agent.py\, including explicit \None\-shadowing on nested child turns.

## Attribution & Lineage

This PR serves as the superseding candidate incorporating:
- Source implementation from @JoaoMarcos44 (#97158)
- Turn-lease cleanup repair from @kshitijk4poor (#97709)
- Generational epoch engine, policy auto-reset advancement, ContextVar nested shadowing, and test suite by @StanleyStetson

## Validation

- **21 unit and integration tests** in \	ests/agent/test_declared_conversation_scope.py\ (including \/new\ rotation, policy auto-resets, ABA prevention, persistence reloads, nested turn shadowing, and wire transports).
- **56 tests** across related suites (\	est_prompt_cache_scope.py\, \	est_48031_model_switch_after_auto_reset.py\) passing 100% green.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] 🚀 New feature (non-breaking change which adds functionality)
- [ ] 🔒 Security fix
- [ ] 📚 Documentation update
- [ ] 🎨 Refactoring (no functional changes)
- [x] 🧪 Tests

### DIFF
--- a/agent/agent_init.py
+++ b/agent/agent_init.py
@@ -597,6 +597,7 @@ def init_agent(
     chat_type: str = None,
     thread_id: str = None,
     gateway_session_key: str = None,
+    gateway_conversation_epoch: int = 1,
     skip_context_files: bool = False,
     load_soul_identity: bool = False,
     skip_memory: bool = False,
@@ -684,6 +685,7 @@ def init_agent(
     agent._chat_type = chat_type
     agent._thread_id = thread_id
     agent._gateway_session_key = gateway_session_key  # Stable per-chat key (e.g. agent:main:telegram:dm:123)
+    agent._gateway_conversation_epoch = gateway_conversation_epoch
     # Pluggable print function — CLI replaces this with _cprint so that
     # raw ANSI status lines are routed through prompt_toolkit's renderer
     # instead of going directly to stdout where patch_stdout's StdoutProxy
--- a/agent/portal_tags.py
+++ b/agent/portal_tags.py
@@ -54,6 +54,30 @@ from typing import List, Optional
 _conversation_id: ContextVar[Optional[str]] = ContextVar(
     "nous_portal_conversation_id", default=None
 )
+_affinity_scope: ContextVar[Optional[str]] = ContextVar(
+    "hermes_affinity_scope", default=None
+)
+
+
+def set_affinity_scope(scope: Optional[str]):
+    """Publish the declared routing/affinity scope for this turn.
+
+    Returns the ContextVar token; pair with :func:`reset_affinity_scope`.
+    """
+    return _affinity_scope.set(scope or None)
+
+
+def reset_affinity_scope(token) -> None:
+    """Restore the previous affinity scope (pair with ``set_affinity_scope``)."""
+    try:
+        _affinity_scope.reset(token)
+    except Exception:
+        _affinity_scope.set(None)
+
+
+def get_affinity_scope() -> Optional[str]:
+    """Return the declared routing/affinity scope, or ``None`` when unset."""
+    return _affinity_scope.get()
 
 
 def set_conversation_context(conversation_id: Optional[str]):
--- a/agent/prompt_cache_scope.py
+++ b/agent/prompt_cache_scope.py
@@ -28,6 +28,37 @@ intentionally different — do not "deduplicate" them.
   timestamp is stripped later by ``_cache_scope_from_session_id`` exactly as
   before.
 
+A host that mints one physical ``session_id`` per RESPONSE (Hermes Studio's
+group chat, and ``POST /v1/responses`` with client-managed history, which
+mints ``str(uuid4())`` per request) re-keys every conversation-affinity hint
+Hermes sends — ``prompt_cache_key`` on both OpenAI-wire transports, plus the
+OpenRouter/Nous sticky ``session_id`` and xAI's ``x-grok-conv-id`` through
+``portal_tags`` (issue #96811). Those rows carry no lineage, so the walk
+above correctly returns the physical id and the scope moves every reply.
+
+Hermes must not infer the logical conversation from the id's SYNTAX (that
+rule collides independent client-supplied ids and merges Studio members
+truncated past its 96-character boundary — the #79017 failure class). The
+host has to declare it, and one carrier already means exactly that:
+``gateway_session_key`` — the "stable per-chat key" (``agent:main:telegram:
+dm:123``) built by ``gateway.session.build_session_key`` from the
+``X-Hermes-Session-Key`` header, which branching deliberately does NOT key
+off. Together with ``gateway_conversation_epoch`` (incremented on ``/new`` /
+``reset_session()``), ``declared_conversation_scope()`` consumes it, and it
+wins over the lineage walk because it is stable across per-response ids while
+rotating on ``/new``. Two boundaries it must not cross:
+
+- explicit fork children (``/branch``, delegate subagents, tool children)
+  share their parent's chat key but are separate conversations — the row's
+  fork markers keep them on their own scope (#79161);
+- background-review forks run on a clone of the live runtime, so they are
+  excluded by ``_persist_disabled`` for the same reason.
+
+The declared key is hashed into ``gwk_<sha256[:24]>`` before it becomes a
+scope: unlike a session id it embeds platform/chat/user identifiers, and
+this value leaves the process verbatim as OpenRouter's sticky ``session_id``
+and xAI's ``x-grok-conv-id``.
+
 The resolution is memoized per (agent, session_id): the lineage walk runs
 once per transcript segment — NOT per API call — and re-runs only when
 rotation actually changes ``agent.session_id`` (per the no-DB-on-the-hot-path
@@ -35,12 +66,15 @@ constraint recorded on #79017). Default installs compact in place and never
 rotate, so they hit the memo forever and behave byte-identically to before.
 """
 
+import hashlib
 import logging
 from typing import Any, Optional
 
 logger = logging.getLogger(__name__)
 
 _MEMO_ATTR = "_prompt_cache_scope_memo"
+# Namespace for a scope resolved from a host-declared conversation key.
+_DECLARED_SCOPE_PREFIX = "gwk_"
 
 
 def _lineage_root(session_id: str, session_db: Any) -> Optional[str]:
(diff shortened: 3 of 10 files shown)


StanleyStetson:
### Context & Architecture

This PR unifies and finalizes the solution for the conversation-affinity caching issue described in #96811, building on the initial work from #97158 and #97709.

#### Problem & Prior Blockers
1. **Session Churn**: Stateless/per-response transports (Hermes Studio Group Chat, `POST /v1/responses`) generate a new physical `session_id` per turn, invalidating upstream prompt caches (`prompt_cache_key`, sticky `session_id`, `x-grok-conv-id`).
2. **Cache Isolation on Reset**: Keying affinity solely off `gateway_session_key` persisted across `/new` (`reset_session()`), breaking cache isolation between distinct conversations (#79017 / #86733).

#### Implemented Solution
- **Generational Scope (`conversation_epoch`)**: `gateway_session_key` is paired with an incrementing `conversation_epoch` and hashed into `gwk_<sha256[:24]>`. Turns within the same conversation reuse the warm cache prefix, while `/new` increments the epoch and rotates the affinity key.
- **Explicit Fork Isolation**: `/branch` children, delegate subagents, tool-spawned sessions, `_reset_from` entries, and background review forks (`_persist_disabled`) explicitly bypass declared keys and isolate to physical session scopes.
- **Fail-Closed Resolution**: DB lookup errors gracefully fallback to physical session IDs (`declared_conversation_scope_safe`).
- **ContextVar Propagation**: `_affinity_scope` is managed via task-local `ContextVar` with safe `try/finally` lifecycle handling in `run_agent.py` and propagation to OpenAI, Codex, OpenRouter, and Nous provider profiles.
- **Test Coverage**: Added 16 unit and integration tests in `tests/agent/test_declared_conversation_scope.py` validating scope stability, epoch rotation, fork isolation, and wire transport invariants.

alt-glitch:

Related: #97709 and #97158 pursue the same host-declared affinity scope. This PR adds conversation-epoch propagation and /new rotation, so it is a competing broader implementation rather than a duplicate.

andrexibiza (commented):
Reviewed exact head `e347fe5436fc8a33d19382217177af6e4fd4b0f1` against its exact parent/current `main@4209d371aa1bb8840ce8447555bdd863a1a96c38`. This is a materially stronger candidate than #97158/#97709: the added `conversation_epoch` directly answers the previously demonstrated `/new` lifetime bug, the fork/background-review exclusions remain intact, and the provider ContextVar propagation is covered end-to-end. Exact-head CI is also genuinely green: CI `33279919278`, Docker `33279918736`, and Nix `33279918716` all completed successfully; the CI run includes Python tests, e2e, Windows/macOS, blocking ruff, Windows footguns, attribution, and `All required checks pass`.

I found one remaining generation bug in the other reset path, plus two landing/topology gates.

## P1 — idle/daily policy resets do not advance the conversation generation, and can ABA back to epoch 1

The new generation is advanced only in the explicit `SessionStore.reset_session()` path:

- `gateway/session.py` adds `SessionEntry.conversation_epoch = 1` and persists it;
- the `reset_session()` hunk constructs the successor with `old_entry.conversation_epoch + 1`;
- `declared_conversation_scope()` then hashes `(gateway_session_key, conversation_epoch)`.

That fixes the exact `/new` defect raised on #97158/#97709.

But `SessionStore.get_or_create_session()` has a second logical-conversation reset path. `_should_reset()` returns `idle`/`daily`; when that fires, `get_or_create_session()` ends/removes the predecessor and later constructs a fresh `SessionEntry(... was_auto_reset=True, auto_reset_reason=..., prev_session_id=...)`. That successor construction does **not** carry `conversation_epoch`, so this PR's dataclass default silently assigns `1`.

That produces two bad reachable histories under one stable gateway key `K`:

```text
conversation A: epoch 1 -> idle/daily reset -> conversation B: epoch 1
                                  same gwk(K), despite a fresh conversation

conversation A: epoch 1 -> /new -> conversation B: epoch 2
conversation B: epoch 2 -> idle/daily reset -> conversation C: epoch 1
                                  reuses A's old generation (ABA)
```

This is not just a cache-performance preference. Main already treats the policy reset as a conversation boundary: `tests/gateway/test_48031_model_switch_after_auto_reset.py` explicitly covers the first message after an `idle/daily/suspended auto-reset` and requires conversation-scope cleanup, while #79017/#86733's accepted cache-scope contract is stable across continuation/compression but fresh across a new conversation.

Required repair: make the epoch a monotonic successor generation for **every** path that replaces one logical conversation with another, not a special field increment inside only `/new`. The policy-reset constructor should derive the successor from the predecessor's generation before the old entry is discarded. Please pin at least these negative controls on a real `SessionStore`:

1. epoch 1 -> idle/daily reset -> epoch 2 and a different `gwk_` scope;
2. epoch 1 -> `/new` -> epoch 2 -> idle/daily reset -> epoch 3, never back to 1;
3. persistence/reload preserves the current epoch before the next reset, so restart cannot recreate an old affinity generation.

The class rule here is the same one the previous `/new` review exposed: a routing coordinate may stay stable, but conversation affinity must be fenced by a monotonic generation that cannot roll back or be reused.

## Landing blocker — preserve the #97158/#97709 origin chain instead of flattening it

This PR is not an independent implementation. Its FILE-LIST contains all seven files from #97158 and #97709 (`portal_tags.py`, `prompt_cache_scope.py`, `hermes_state.py`, both provider plugins, `run_agent.py`, and the declared-scope test), then adds the epoch projection in four more files.

That history matters because #97709 deliberately preserved JoaoMarcos44's original authored implementation commit and kept kshitijk4poor's `affinity_token` cleanup repair as a separate authored commit. The #97709 thread then identified the missing per-conversation epoch; this PR implements exactly that next refinement. Current #98170 collapses the whole chain into one StanleyStetson-authored commit even though its discussion correctly says it builds on both prior PRs.

Please preserve that lineage in the landing object: source implementation from JoaoMarcos44 (#97158), cleanup repair from kshitijk4poor (#97709), then this epoch refinement from StanleyStetson. If history has to be recomposed on current main, the equivalent provenance must still be explicit in commit/PR attribution. #98170 is best understood as the broader **superseding** candidate once the generation bug above is fixed; #97158 is the source implementation and #97709 is the salvage/cleanup refinement, not disposable duplicate work.

Because all three collide on the same seven files, only one should land. Once the final object is correct, the other two can be closed as superseded with their contributions named rather than silently orphaned.

## Structural gate — this adds new authority to three existing godfiles

The new behavior lands inside `gateway/run.py` (>29k lines), `hermes_state.py` (>13k), and `run_agent.py` (>9k). That violates the repository's 2K invariant for modified files. More importantly for this particular change, `gateway/run.py` now reads `session_store._entries` directly to project `conversation_epoch` into `TurnContext`, so the generation lookup itself is outside the store's owned abstraction/lock boundary.

The narrow shape I would land is a bounded session-affinity/session-generation seam that owns: successor generation, safe lookup, fork classification, and turn projection. `gateway/run.py` should consume a public value, not reach into the private `_entries` map; `hermes_state.py` should not grow another public classifier in place if the session-lineage seam can own it. That extraction also gives the idle/daily fix one authoritative place instead of duplicating generation behavior across reset call sites.

## Interlock / merge order

- #96811 is the live P0 this candidate is intended to close.
- #79017 / merged #86733 define the existing cache-scope boundary: continuation/compression inherits; new conversations/forks isolate.
- #97158 is the original host-declared-affinity implementation by JoaoMarcos44.
- #97709 is the salvage by kshitijk4poor that preserves that authorship, fixes the cleanup regression, and explicitly identifies the per-conversation epoch as the remaining design requirement.
- #98170 is therefore a complementary refinement **and** the natural superseder, provided it closes the policy-reset generation path and lands with the origin chain intact.

The exact-head green receipts are strong, and the epoch direction is the right answer to the earlier `/new` blocker. The remaining runtime defect is narrow but important: generation must advance at every conversation replacement, otherwise the new proof object can roll backward and reuse a prior affinity identity. Fix that, preserve the contributor lineage, extract the owning seam, and this becomes a much cleaner closure for #96811. 🚀

StanleyStetson:
Thank you @andrexibiza for the thorough review and clear guidance!

All three items have been implemented, verified, and pushed to the branch (`4e48356d95`):

### 1. P1: Monotonic Generation on Idle/Daily Auto-Resets (ABA Prevention)
- In `gateway/session.py::get_or_create_session()`, when `_should_reset()` fires (`idle`, `daily`, or stale recovery), the successor session constructor now derives `conversation_epoch = getattr(predecessor, "conversation_epoch", 1) + 1`.
- Added 3 new negative control / regression tests in `tests/agent/test_declared_conversation_scope.py`:
  1. `test_auto_reset_advances_epoch_and_gwk_scope`: asserts `idle` policy reset bumps epoch 1 to 2 and rotates the `gwk_` scope.
  2. `test_mixed_new_and_auto_reset_never_rolls_back_epoch_aba`: asserts epoch 1 to `/new` (2) to `daily` reset (3), strictly preventing ABA rollback to 1.
  3. `test_persistence_and_reload_preserves_epoch_before_reset`: asserts epoch state survives store reconstruction/restart across cycles.

### 2. SessionStore Abstraction Seam
- Added public `SessionStore.get_conversation_epoch(session_key: str) -> int` method.
- Refactored `gateway/run.py` to consume this public method rather than accessing private `session_store._entries`.

### 3. Provenance & Co-Authorship Lineage
- Explicitly attributed the origin chain in the commit message and PR body, preserving contributions from @JoaoMarcos44 (#97158) and @kshitijk4poor (#97709) with `Co-authored-by:` git trailers.

### Validation
- All **21 tests** in `tests/agent/test_declared_conversation_scope.py` pass.
- All **56 regression tests** across related suites (`test_prompt_cache_scope.py`, `test_48031_model_switch_after_auto_reset.py`, `transports/test_chat_completions.py`) pass 100% green.

kshitijk4poor:
This is a serious attempt at exactly the design gap flagged in the #97158 review (declared key + epoch so the scope survives per-response ids but rotates on /new) — and the epoch direction is right. Closing this particular implementation on three verified defects rather than carrying it as an open P0:

1. **Hot-path regression: it defeats the memoization invariant.** The diff inserts `declared_conversation_scope(agent)` BEFORE the memo lookup in `resolve_prompt_cache_scope()`. On current main `SessionDB` has no public `is_explicit_fork_child`, so your `getattr` fallback runs `db.get_session(sid)` + `_is_explicit_fork_child_row` — a session-row SELECT on **every API call** for every gateway agent (this resolver is called from `turn_context` and `chat_completion_helpers` per request). The module's own docstring mandates no-DB-on-the-hot-path via the per-segment memo; this reintroduces exactly what that design removed. The declared-scope check has to live inside the memoized path.

2. **Epoch coverage misses the host that motivated the bug.** Epoch is wired only through `gateway/run.py`'s TurnContext and `SessionStore.reset_session`. The `/v1/responses` client-managed-history boundary and Hermes Studio group chat — the two per-response hosts from #96811 — get the default epoch 1 forever, and the `epoch != 1` special-case means their carrier is the bare chat key, i.e. the same `/new`-lifetime problem the #97158 review reproduced remains for any reset semantics those hosts express. Out-of-turn compaction publication (covered in #97709) also isn't wired here, so aux/compaction traffic falls back to the physical id mid-conversation.

3. **Attribution.** The core of `declared_conversation_scope`, the `gwk_` scheme, fork/background-review exclusions, and much of the docstring text are lifted from @JoaoMarcos44's #97158 in a single commit under your authorship. Salvage/extension is welcome, but contributor commits must be preserved (cherry-pick) — see the repo's contributor-credit policy.

The epoch concept itself is now part of the design discussion on #97158, where the lifecycle question is being settled; the eventual revision will land through that thread (with #97709 as the salvage vehicle). Thanks — the ABA-monotonicity tests here are a genuinely useful checklist for that revision.

StanleyStetson:
@kshitijk4poor  Thank you for the detailed review and for highlighting
the gaps — I accept all three points and will address them.

On the authorship point: my goal was to contribute and
help close the issue faster, never to claim someone
else's work. The commit structure came out sloppy —
I'll fix it.

Continuing the discussion in #97158 as suggested.

PR #65216, opened 2026-07-15
fix(telegram): create System topic on auto-threaded first /topic (#65202)

wesleysimplicio:
## Summary

Running `/topic` from Telegram's **All Messages** view enables multi-session
topic mode but does not create the managed `System` topic. Telegram
auto-creates a topic for the `/topic` message itself, so the command reaches
Hermes with a `thread_id` even on the very first activation.
`_handle_topic_command` gated System-topic creation on `not source.thread_id`,
so that auto-threaded first activation skipped `_ensure_telegram_system_topic`
entirely. Sending `/topic` through the Bot API *without* a topic works, which
is exactly the `thread_id`-present vs -absent split the reporter observed.

## Changes

- `gateway/slash_commands.py` — in `_handle_topic_command`, read
  `is_telegram_topic_mode_enabled` **before** enabling topic mode and gate
  `_ensure_telegram_system_topic` on whether topic mode was *already* enabled
  (the correct "first activation" signal), instead of on `source.thread_id`.
  The read is wrapped defensively (defaults to "not enabled" on error) and runs
  on the existing async DB facade. Because `_ensure_telegram_system_topic`
  creates a fresh `System` forum topic and is **not** idempotent, gating on
  first-activation also prevents re-creating a duplicate `System` topic on a
  repeat `/topic`.
- `tests/gateway/test_telegram_topic_mode.py` — regression coverage:
  - `test_topic_first_activation_creates_system_topic_when_auto_threaded`:
    first `/topic` arriving with a `thread_id` (auto-created topic) must still
    create the System topic. **Fails on `main`** (System topic never created).
  - `test_topic_repeat_activation_does_not_recreate_system_topic`: a second
    `/topic` once mode is enabled must not create another System topic.

## Validation

| Command | Result |
| --- | --- |
| `pytest .../test_topic_first_activation_creates_system_topic_when_auto_threaded` (on `main`, no fix) | **FAIL** — `_ensure_telegram_system_topic` awaited 0 times |
| `pytest -k "system_topic or recreate_system"` (with fix) | 3 passed |
| `pytest tests/gateway/test_telegram_topic_mode.py` | 47 passed |
| `ruff check gateway/slash_commands.py tests/gateway/test_telegram_topic_mode.py` | All checks passed |

## Behavior preserved

- `/topic` from root when mode disabled → still creates the System topic.
- `/topic` inside an existing topic when mode already enabled → still a no-op
  for System-topic creation (was `not thread_id`, now `already-enabled`; same
  outcome).

[...]

PR #98172, opened 2026-08-29
fix(gateway): stop creating the unused Telegram System topic on /topic activation

### LINKED ISSUE #98066: `/topic` creates a "System" topic that nothing writes to and nothing tracks
**Version checked:** `main` @ `3f36c87e1e` (2026-08-29)

### Summary

Enabling multi-session DM mode with `/topic` creates a Telegram forum topic named
"System", posts one intro message into it, and pins it
(`_ensure_telegram_system_topic`, `gateway/run.py:23788`).

Nothing ever uses that topic again. It looks like a leftover from a design that
was superseded the day after it landed.

### Evidence

**1. Nothing ever writes to it.** The string `"System"` appears exactly once in
the whole codebase — the `create_topic` call itself:

```
$ git grep -n '"System"' -- 'gateway/*' 'plugins/*' 'hermes_plugins/*'
gateway/run.py:23798:  thread_id = await create_topic(int(source.chat_id), "System")
```

After the intro message and the pin, no code path targets it. Watch-pattern
notifications route to the thread of the originating event, not to a fixed lane.

**2. Its `thread_id` is never persisted.** The `telegram_dm_topic_mode` table has
`intro_message_id` and `pinned_message_id` columns, and no code writes them:

```
$ git grep -n "intro_message_id" -- '*.py'
hermes_state.py:14410:   intro_message_id TEXT,     # CREATE TABLE only
```

So the gateway cannot find the topic it created, cannot tell whether it still
exists, and cannot notice when a user deletes it.

**3. The documented system lane is a different place.** `/topic help` says
(`gateway/run.py:24292`):

> 3. The root DM becomes a system lobby — send /topic, /status, /help, /usage
>    there. Normal prompts go in a topic.

The root DM is the lobby. The "System" topic is a second, competing lane that the
help text never mentions.

### How it got here

- **2026-05-02** — `25065283b3` *"fix: improve telegram topic mode setup"* adds
  the System topic: create, intro, pin.
- **2026-05-03**, one day later — `d35efb9898` adds the `/topic help` text that
  assigns the system-lane role to the root DM instead. The creation code is left
  in place.
- **2026-07-29** — `6b81590c55` *"test: prune low-value tests suite-wide
  (wave 1)"* deletes `test_topic_root_command_creates_and_pins_system_topic`.
  The test that pinned the behaviour is gone; the behaviour stays, now uncovered.

### Impact

Small but not zero:

- Every `/topic` activation creates a topic the user has no reason to use, and
  which the help text never explains.
- Deleting it (easy to do by accident — it looks like clutter) is invisible to
  the gateway. Re-running `/topic` from inside a topic will not recreate it,
  because the call is guarded by `if not source.thread_id`, and a failed
  `_create_dm_topic` returns `None` while only logging at `debug`. From the
  outside it looks like the bot silently lost a feature.
- Two dead schema columns are kept alive by it.

### Suggested fix

Remove `_ensure_telegram_system_topic` and its call site in
`gateway/slash_commands.py`; the root DM already plays the role the help text
documents. That also lets the two unused columns go.

If the separate System lane *is* intended, the opposite fix is to give it the
role it never had: persist its `thread_id`, and route system-level output
(status, notifications, warnings) into it.

Happy to send a PR for either direction — just say which one you'd take.

---

**Related:** #65202 reports the opposite symptom — `/topic` *failing* to create
the System topic when Telegram auto-threads the command. That is the same
`if not source.thread_id` guard described under Impact, seen from the other side.
Worth deciding together: #65202 assumes the lane should exist, this one asks
whether it should. If it should, that issue is the concrete case the
persist-and-route fix would have to cover.

liuhao1024:
## What does this PR do?

`/topic` activation mints a Telegram forum topic named "System", posts one intro message into it, and pins it. Nothing ever uses that topic again — this PR removes the leftover creation path.

Three facts pin it down as a leftover (per #98066):

1. **Nothing ever writes to it.** The string `"System"` appears exactly once in the whole codebase — the `create_topic` call itself. Watch-pattern notifications route to the thread of the originating event; nothing targets a fixed lane.
2. **Its `thread_id` is never persisted.** The `telegram_dm_topic_mode` table has `intro_message_id` / `pinned_message_id` columns that exist in `CREATE TABLE` only — no code writes them. The gateway cannot find the topic it created, cannot tell whether it still exists, and cannot notice when a user deletes it.
3. **The documented system lane is a different place.** `/topic help` says "The root DM becomes a system lobby — send /topic, /status, /help, /usage there." The root DM is the lobby; the "System" topic is a second, competing lane the help text never mentions.

Git history shows why: `25065283b3` ("fix: improve telegram topic mode setup", 2026-05-02) added the System topic; one day later `d35efb9898` introduced the `/topic help` text that assigns the system-lane role to the root DM instead, leaving the creation code in place. The behavior-pin test for it was later deleted in `6b81590c55` (low-value test prune), so the behavior has been uncovered since.

Removing the creation call is the right fix rather than persisting the topic: the documented design (root DM as lobby) already covers the system-lane role, and keeping a second lane alive would require new persistence and routing that nothing asks for.

## Related Issue

Fixes #98066

## Type of Change

## Changes Made

- `gateway/slash_commands.py` — removed the `if not source.thread_id: await self._ensure_telegram_system_topic(source)` call after `enable_telegram_topic_mode` succeeds
- `gateway/run.py` — removed the now-unreferenced `_ensure_telegram_system_topic` helper (create + intro send + pin, all debug-swallow failures)
- `tests/gateway/test_telegram_topic_mode.py` — added `test_topic_activation_does_not_create_system_topic` pinning that `/topic` activation never calls `_create_dm_topic`

## How to Test

1. `python -m pytest tests/gateway/test_telegram_topic_mode.py -q` — Observed result: 20 passed (was 19 before this change; the new regression test passes with the creation path removed).
2. Red-check: re-adding the `_ensure_telegram_system_topic` call makes the new test fail (`AttributeError` / `_create_dm_topic` called), confirming it actually pins the removal.
3. `python -m pytest tests/gateway/test_slash_access_dispatch.py tests/gateway/test_telegram_forum_commands.py tests/gateway/test_dm_topics.py tests/gateway/test_base_topic_sessions.py -q` — Observed result: 26 passed (adjacent slash-command and topic surfaces unaffected).
4. `grep -rn "_ensure_telegram_system_topic" --include="*.py" .` — Observed result: no references remain.

### Code

- [x] I have read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this is not a duplicate
- [x] I have run `pytest tests/ -q` and all tests pass
- [x] I have added tests for my changes (required for bug fixes, strongly encouraged for features)
- [x] I have tested on my platform: macOS 15 (arm64)

- [x] N/A — documentation update (no user-facing docs mention the System topic; `/topic help` already documents the root-DM lobby)
- [x] N/A — `cli-config.yaml.example` (no config keys changed)
- [x] N/A — `CONTRIBUTING.md` / `AGENTS.md` (no architecture or workflow change)
- [x] N/A — cross-platform impact (server-side gateway logic; no OS-specific behavior touched)
- [x] N/A — tool descriptions/schemas (no tool behavior changed)

## Screenshots / Logs

```text
$ python -m pytest tests/gateway/test_telegram_topic_mode.py -q
20 passed, 1 warning in 2.70s

$ python -m pytest tests/gateway/test_telegram_topic_mode.py::test_topic_activation_does_not_create_system_topic -q   # with the call temporarily re-added
1 failed  (AttributeError: _ensure_telegram_system_topic ... / _create_dm_topic called)
```

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -23785,47 +23785,6 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
             "allows_users_to_create_topics": _field("allows_users_to_create_topics"),
         }
 
-    async def _ensure_telegram_system_topic(self, source: SessionSource) -> None:
-        """Create/pin the managed System topic after /topic activation when possible."""
-        adapter = self._adapter_for_source(source)
-        if adapter is None or not source.chat_id:
-            return
-
-        thread_id = None
-        create_topic = getattr(adapter, "_create_dm_topic", None)
-        if callable(create_topic):
-            try:
-                thread_id = await create_topic(int(source.chat_id), "System")
-            except Exception:
-                logger.debug("Failed to create Telegram System topic", exc_info=True)
-        if not thread_id:
-            return
-
-        message_id = None
-        try:
-            send_result = await adapter.send(
-                source.chat_id,
-                "System topic for Hermes commands and status.",
-                metadata={"thread_id": str(thread_id)},
-            )
-            message_id = getattr(send_result, "message_id", None)
-        except Exception:
-            logger.debug("Failed to send Telegram System topic intro", exc_info=True)
-        if not message_id:
-            return
-
-        bot = getattr(adapter, "_bot", None)
-        if bot is None or not hasattr(bot, "pin_chat_message"):
-            return
-        try:
-            await bot.pin_chat_message(
-                chat_id=int(source.chat_id),
-                message_id=int(message_id),
-                disable_notification=True,
-            )
-        except Exception:
-            logger.debug("Failed to pin Telegram System topic intro", exc_info=True)
-
     async def _send_telegram_topic_setup_image(self, source: SessionSource) -> None:
         """Send the bundled BotFather Threads Settings screenshot when available."""
         adapter = self._adapter_for_source(source)
--- a/gateway/slash_commands.py
+++ b/gateway/slash_commands.py
@@ -4880,9 +4880,6 @@ class GatewaySlashCommandsMixin:
             logger.exception("Failed to enable Telegram topic mode")
             return t("gateway.topic.enable_failed", error=exc)
 
-        if not source.thread_id:
-            await self._ensure_telegram_system_topic(source)
-
         if source.thread_id:
             try:
                 binding = await self._session_db.get_telegram_topic_binding(
--- a/tests/gateway/test_telegram_topic_mode.py
+++ b/tests/gateway/test_telegram_topic_mode.py
@@ -830,3 +830,36 @@ def test_get_telegram_topic_binding_by_session_returns_binding(tmp_path):
 # Test for session-split thread_id recovery (issue #27166)
 # ---------------------------------------------------------------------------
 
+
+# ---------------------------------------------------------------------------
+# Regression: /topic activation must not create the "System" forum topic
+# (issue #98066)
+# ---------------------------------------------------------------------------
+
+@pytest.mark.asyncio
+async def test_topic_activation_does_not_create_system_topic(tmp_path, monkeypatch):
+    """/topic activation must not mint a "System" forum topic.
+
+    The root DM is the documented system lobby (/topic help), and the
+    "System" topic's thread id was never persisted — no code path ever
+    read from it, so it only ever produced an unexplained pinned topic
+    that silently survived deletion (#98066).
+    """
+    import gateway.run as gateway_run
+
+    session_db = SessionDB(db_path=tmp_path / "state.db")
+    runner = _make_runner(session_db=session_db)
+    runner._run_agent = AsyncMock(
+        side_effect=AssertionError("root /topic activation must not enter the agent loop")
+    )
+
+    monkeypatch.setattr(
+        gateway_run, "_resolve_runtime_agent_kwargs", lambda: {"api_key": "***"}
+    )
+
+    result = await runner._handle_message(_make_event("/topic"))
+
+    assert "Telegram multi-session topics are enabled" in result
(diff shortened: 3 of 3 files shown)


alt-glitch:

Related: #65216 preserves System-topic initialization for the auto-threaded /topic path, while this PR removes the untracked root-activation System topic. Please reconcile the intended Telegram topic model before merging either direction.

liuhao1024:
Thanks for flagging the overlap — happy to reconcile.

To make the delta precise: this PR removes the **root-activation** branch of `/topic` (the `if not source.thread_id:` call site in `_handle_topic_command`) plus the then-unreferenced `_ensure_telegram_system_topic` in `gateway/run.py`. #65216 works on the **auto-threaded first-activation** path (a different call site, earlier in the same handler) and adds idempotency gating there — but its auto-threaded branch still calls the same `_ensure_telegram_system_topic`, so merge order matters:

- **If the intended model is "#65216's System topic, auto-threaded only"**: the two PRs are compatible in spirit — root activation still mints a topic whose thread_id is never persisted (`telegram_dm_topic_mode`'s `intro_message_id`/`pinned_message_id` columns have no writer in either PR) — but I would rebase this to keep `_ensure_telegram_system_topic` alive for the auto-threaded call site.
- **If the intended model is "System topic across both activation paths"**: this PR should be dropped in favor of #65216 extending its gating to the root path.

Both directions are fine by me — deferring to the maintainers on which topic model is intended, and I'll rebase or withdraw accordingly.

PR #98175, opened 2026-08-29
fix(buzz): bound WebSocket read idle time to force reconnect on silent relays

### LINKED ISSUE #98097: Buzz adapter: WebSocket silently enters CLOSE_WAIT after long idle; gateway stays 'connected' forever; inbound stops
Title:
Buzz adapter: WebSocket silently enters CLOSE_WAIT after long idle; gateway stays "connected" forever; inbound stops

Body:
## Environment

- Hermes Agent v0.20.4 (tag `v2026.8.18`, commit `e624e9f`)
- Linux (Ubuntu 24.04, kernel 6.8.0-138), running as unprivileged user `hermes` under user-systemd
- `plugins/platforms/buzz` at the same tag
- Self-hosted Buzz relay v0.5.2 (image `ghcr.io/block/buzz:sha-3e48f1b`) at `https://buzz.ubikitos.com`, behind Caddy → Cloudflare edge
- Model: `nous` provider, `tencent/hy3:free`, authenticated via OAuth device code

## Symptom

After ~7 hours of continuous uptime with the gateway connected to the relay and answering messages correctly, the gateway silently stops receiving new inbound messages. The user can see their outgoing message reach the channel in the Mac Buzz Desktop client, but CHIEF never replies and `gateway_state.json` still shows:

```json
"platforms": {
  "buzz": {
    "state": "connected",
    "error_code": null,
    "error_message": null,
    "updated_at": "2026-08-29T12:25:22.546904+00:00",
    "writer_pid": 786246
  }
}
```

A full `systemctl --user restart hermes-gateway-chief` recovers immediately and the next inbound message is processed within `poll_interval` (4s) plus model latency.

## Reproduction (deterministic)

1. Stand up a self-hosted Buzz relay (or use the public test relay behind Cloudflare — Cloudflare's 100s idle timeout is the trigger).
2. Configure a Buzz Hermes gateway with `poll_interval: 4`, `channels: [general]`.
3. Send a message; confirm a successful reply.
4. Leave the gateway idle (no further messages) for **~6–7 hours** (Cloudflare's typical idle LB close window).
5. Send a message from a different client. The relay accepts it; the user sees it in the channel. The gateway does NOT pick it up.
6. `cat /home/hermes/.hermes/profiles/chief/gateway_state.json` still shows `state: "connected"`, `error_code: null`.

## Kernel-level evidence

`ss -tnp` on the gateway's host during the dead state shows:

```
ESTAB     147.79.118.32:37178  147.79.118.32:443     (an unrelated ESTAB)
ESTAB     [2a02:…]:36194        2606:4700:…:443       (new connection, alive)
CLOSE-WAIT [2a02:…]:42354       2606:4700:…:443      (the buzz.ubikitos.com WebSocket — DEAD)
```

The Cloudflare edge sent a FIN; the kernel marked the socket `CLOSE_WAIT`; the `websockets` library did not surface this as a connection failure, so the adapter's read loop is still parked on it and `connect()` was never re-entered. The reconnect watcher (`gateway/run.py:4042`) only fires on `connect()` failure, so the dead socket is invisible to it.

## Why this is a class of bug, not a one-off

Related issues that share the same root shape (silently-stuck connection, no error surface, only a restart fixes it):

- #69314 — Telegram gateway, same CLOSE_WAIT pattern, healthy HTTP proxy. Closed as duplicate of #68983 (`fix(telegram): recover exhausted request pool`).
- #62175 — Dashboard leaking CLOSE_WAIT sockets to Nous cloud; the #18766 keepalive fix was insufficient at HEAD.
- #76385 — "Buzz gateway can stay connected while the agent appears offline" — different surface (presence display, not message delivery), but the same diagnostic pattern of a long-lived connection not noticing that the other end is gone.

In our case the difference from #76385 is the failure surface: it's message delivery that stops, not just the presence indicator. The mechanism (no read-side liveness check, no pong/heartbeat timeout, no proactive `CLOSE_WAIT` detection) is the same class.

## Expected

The Buzz adapter should detect that a previously-opened WebSocket has become a zombie (peer-side close, no events for `>` heartbeat budget, or `CLOSE_WAIT` / `FIN_WAIT` from the kernel) and trigger `is_reconnect=True` to cycle the connection without a full gateway restart. A user-settable heartbeat budget (default e.g. 90s, three times the typical 30s relay keepalive) would be the minimum useful change.

## Workaround in production

For now I am running `systemctl --user restart hermes-gateway-chief` whenever this happens (manually detected by monitoring the gateway's outbound socket states). I have NOT installed a watchdog because the right long-term fix is upstream and I don't want to paper over the real issue.

liuhao1024:
## What does this PR do?

The Buzz adapter's WebSocket read loop had no read-side liveness bound. When a relay-side close is never surfaced by the transport — observed in #98097 as a socket stuck in CLOSE_WAIT behind Cloudflare's ~100s idle window — the read loop stays parked on the frame iterator forever: inbound messages stop, `gateway_state.json` still reports `connected` with no error code, and only a full gateway restart recovers. The library keepalive (`ping_interval=20`, `ping_timeout=20`) is the first line of defense, but this change adds a last-resort bound: each frame read now waits at most `_WS_READ_IDLE_TIMEOUT` (300s). On expiry the loop raises into its existing disconnect/reconnect path, which re-authenticates (NIP-42), re-subscribes with per-channel `since` filters intact, and preserves the bounded backoff. A healthy idle relay costs one reconnect every 5 minutes at worst — the per-channel `since` resume makes that lossless.

## Related Issue

Fixes #98097

## Type of Change

## Changes Made

- `plugins/platforms/buzz/adapter.py`: added `_WS_READ_IDLE_TIMEOUT = 300.0` next to the other WS constants; replaced the bare `async for raw in websocket` in `_websocket_loop` with an explicit `__aiter__/__anext__` loop wrapped in `asyncio.wait_for(..., timeout=_WS_READ_IDLE_TIMEOUT)`, converting a read timeout into the same `ConnectionError` shape the loop already handles (CLOSED frames, auth failures) so the existing backoff/reconnect wiring is reused unchanged. Frame dispatch semantics are identical: same JSON parsing, same EVENT/CLOSED/NOTICE routing, `StopAsyncIteration` still exits the connection cleanly.

## How to Test

1. Run the WebSocket transport tests: `.venv/bin/python -m pytest tests/gateway/test_buzz_websocket.py -q` — Observed result: 6 passed (4 pre-existing + the 2 new regression tests).
2. Run the adapter suite: `.venv/bin/python -m pytest tests/gateway/test_buzz_adapter.py -q` — Observed result: 23 passed.
3. New regression test `test_websocket_loop_reconnects_when_read_goes_silent`: a scripted connection whose `__anext__` never yields and never raises (the CLOSE_WAIT shape, with the idle timeout patched to 0.05s) should force a second `connect()` and close the silent connection first — observed passing, with the `went silent` warning logged.
4. New regression test `test_websocket_loop_dispatches_frames_and_closes_cleanly`: a healthy scripted frame dispatches through `_handle_event` and a clean server-side close (`StopAsyncIteration`) exits the `async with` block before reconnect — observed passing.

### Code

- [x] I've run `pytest tests/ -q` and all tests pass (buzz suites: tests/gateway/test_buzz_websocket.py 6 passed, tests/gateway/test_buzz_adapter.py 23 passed)
- [x] I've tested on my platform: macOS 26.5 (arm64)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (the constant carries an explanatory comment)
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — or N/A (no new config keys)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (pure asyncio, no platform-specific code)

### DIFF
--- a/plugins/platforms/buzz/adapter.py
+++ b/plugins/platforms/buzz/adapter.py
@@ -103,6 +103,13 @@ _CLI_TIMEOUT = 30.0
 # WebSocket transport (NIP-42 authenticated Nostr subscription).
 # kind 44100 is Buzz's channel-membership event — used for live DM discovery.
 _WS_AUTH_TIMEOUT = 20.0
+# Last-resort bound on how long the read loop may wait for a frame. The
+# library keepalive (ping_interval/ping_timeout below) should catch a dead
+# relay first, but a relay-side close the transport never surfaces (observed
+# as a CLOSE_WAIT socket with the loop parked on recv, #98097) leaves the
+# gateway "connected" while inbound stops; this timeout forces the normal
+# reconnect path instead.
+_WS_READ_IDLE_TIMEOUT = 300.0
 _WS_MAX_MESSAGE_BYTES = 2_000_000
 _WS_MEMBERSHIP_KIND = 44100
 _WS_MEMBERSHIP_SUB_ID = "hermes-buzz-membership"
@@ -864,7 +871,20 @@ class BuzzAdapter(BasePlatformAdapter):
                         if self._ws_ready is not None:
                             self._ws_ready.set()
                         backoff = 1.0
-                        async for raw in websocket:
+                        frame_iter = websocket.__aiter__()
+                        while True:
+                            try:
+                                raw = await asyncio.wait_for(
+                                    frame_iter.__anext__(),
+                                    timeout=_WS_READ_IDLE_TIMEOUT,
+                                )
+                            except StopAsyncIteration:
+                                break
+                            except asyncio.TimeoutError:
+                                raise ConnectionError(
+                                    f"no WebSocket frame for {_WS_READ_IDLE_TIMEOUT:.0f}s; "
+                                    "assuming the connection went silent"
+                                ) from None
                             try:
                                 message = json.loads(raw)
                             except (ValueError, TypeError):
--- a/tests/gateway/test_buzz_websocket.py
+++ b/tests/gateway/test_buzz_websocket.py
@@ -104,6 +104,134 @@ class _FakeWebSocket:
         self.sent.append(json.loads(raw))
 
 
+# ── _websocket_loop: read-idle watchdog (#98097) ──────────────────────────
+
+
+class _ScriptedWebSocket(_FakeWebSocket):
+    """A connect() target whose event frames come from a scripted behavior.
+
+    The auth handshake is the relay's (inherited from _FakeWebSocket); after
+    it, each ``__anext__`` delegates to ``anext_behavior`` — a coroutine
+    function returning the next raw frame or raising StopAsyncIteration for
+    a clean close.
+    """
+
+    def __init__(self, anext_behavior):
+        super().__init__()
+        self._anext_behavior = anext_behavior
+        self.exited = False
+
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, *exc_info):
+        self.exited = True
+
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self):
+        return await self._anext_behavior()
+
+
+@pytest.mark.asyncio
+async def test_websocket_loop_reconnects_when_read_goes_silent(monkeypatch, caplog):
+    """A relay close the transport never surfaces must not park the loop.
+
+    Reproduces the #98097 shape: a socket stuck in CLOSE_WAIT yields no
+    frame and no error, so without a read-side bound the loop would wait
+    forever while the gateway keeps reporting "connected".
+    """
+    import logging
+
+    adapter = _make_adapter()
+    monkeypatch.setattr(_buzz_mod, "_WS_READ_IDLE_TIMEOUT", 0.05)
+    caplog.set_level(logging.WARNING)
+
+    sockets = []
+
+    async def dead_anext():
+        await asyncio.Event().wait()  # never yields, never raises
+
+    def fake_connect(*args, **kwargs):
+        ws = _ScriptedWebSocket(dead_anext)
+        sockets.append(ws)
+        return ws
+
+    import websockets as _ws_mod
+
+    monkeypatch.setattr(_ws_mod, "connect", fake_connect)
+
+    task = asyncio.create_task(adapter._websocket_loop())
+    try:
+        deadline = time.monotonic() + 5.0
+        while len(sockets) < 2 and time.monotonic() < deadline:
+            await asyncio.sleep(0.02)
+    finally:
+        task.cancel()
+        try:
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Well-scoped liveness fix for Buzz relay silent-close (CLOSE_WAIT, #98097). Keeps library ping keepalive as first line, adds last-resort 300s read-idle watchdog that forces normal reconnect instead of parking forever on `recv`.

**What it does**
- `plugins/platforms/buzz/adapter.py:15` adds `_WS_READ_IDLE_TIMEOUT = 300.0` with comment explaining Cloudflare ~100s window and library `ping_interval/ping_timeout=20`.
- `adapter.py:871` refactors `async for raw in websocket:` to manual `frame_iter = websocket.__aiter__()` + `asyncio.wait_for(frame_iter.__anext__(), timeout=...)`; `TimeoutError` raises `ConnectionError("no WebSocket frame for 300s; assuming the connection went silent")` to trigger outer reconnect, `StopAsyncIteration` breaks cleanly.
- Tests `tests/gateway/test_buzz_websocket.py:80` mock `_WS_READ_IDLE_TIMEOUT=0.05`, scripted websocket that never yields — asserts loop reconnects (≥2 sockets), first exited, log contains "went silent"; second test asserts healthy frame dispatch + clean `StopAsyncIteration` handling preserved.

**Non-blocking notes**
- `wait_for` wrapping `__anext__` leaves the underlying `__anext__` task pending on timeout cancellation — `websockets` cancel propagation is normally safe, but ensure outer `async with websocket` cleanup (`self.exited`) still runs before next `connect()` (test asserts it does).
- 300s is intentionally generous vs ping timeout (20+20) to avoid competing with keepalive; no change to ping semantics.
- Second test's second `connect` raises `CancelledError` to terminate loop — test-only harness, not prod behavior.

liuhao1024:
Thanks for the review. On the `wait_for`/`__anext__` cancellation note: the reconnect path always exits the `async with websocket` block before the next `connect()`, and `test_silent_websocket_forces_reconnect` asserts exactly that (first socket `exited` set, ≥2 connects). Agreed the 300s watchdog stays a last resort behind the 20s+20s ping keepalive.

teknium1:
Merged via #99430 — your read-idle watchdog commit was cherry-picked with authorship preserved, regression test included. Fixes #98097.

PR #98176, opened 2026-08-29
fix(kanban): make kanban_block idempotent on already-blocked tasks

vova-superstrata:
## Summary

`kanban_block` returns a hard error when the task is **already blocked** — which happens routinely when a block call is retried (double tool delivery, judge-driven goal-loop retry, orphaned tool result after an interrupted turn). The first call succeeds and blocks the task; the retry then fails with the catch-all:

```
could not block t_X (unknown id or not in running/ready)
```

The worker's run then ends on a spurious failure even though the intended state (task blocked) was fully achieved. The same catch-all also conflates two genuinely different failures — unknown task id vs. task in a non-blockable state — leaving the caller unable to tell which occurred.

Real-world impact from one deployment's tool-call history: 12 of 17 lifetime `kanban_block` errors were this exact double-block pattern (e.g. a task successfully blocked at 20:23, then an identical retry at 20:24 hard-erroring).

## Fix

In `_handle_block`, when `block_task` returns false, disambiguate before erroring:

- **Already blocked** (status `blocked`, or `todo` with an active `block_kind` for dependency blocks) → return `ok: true` with `idempotent: true` and a note; the existing block record is preserved untouched (a retry never overwrites the original reason/kind).
- **Unknown id** → `could not block t_X (unknown id)`.
- **Non-blockable state** (e.g. `done`) → precise error naming the actual status: `task is in status 'done', not running/ready (a task that has already finished cannot be blocked)`.

Caller-contract violations (goal-mode kind restrictions, worker task-scoping) are untouched and still rejected — this only makes the *success* case idempotent.

## Testing

- 5 new regression tests in `tests/tools/test_kanban_tools.py` covering idempotent double-block (both `blocked`-routing and `dependency`→`todo`-routing kinds), preservation of the original block, and the two distinct error paths.
- Red/green verified: with the fix reverted, 3 of the new tests fail on base; with the fix, the full file passes 36/36.
- E2E on a real scratch board (actual DB path via `HERMES_KANBAN_HOME`): all six behaviour cases correct.
- Paired full filtered suite (`-k "block or kanban"`, ~1,700 tests) on clean base vs. patched: byte-identical failure sets — no regressions introduced.

### DIFF
--- a/tests/tools/test_kanban_tools.py
+++ b/tests/tools/test_kanban_tools.py
@@ -228,6 +228,77 @@ def test_block_happy_path(worker_env):
         conn.close()
 
 
+def test_block_already_blocked_is_idempotent(worker_env):
+    """Re-blocking a task whose block already landed (double delivery,
+    judge-driven retry, orphaned tool result) must be a no-op success, not a
+    hard error that ends the worker's run on a spurious failure."""
+    from tools import kanban_tools as kt
+    first = json.loads(kt._handle_block({"reason": "need clarification"}))
+    assert first["ok"] is True
+    second = json.loads(kt._handle_block({"reason": "need clarification"}))
+    assert second["ok"] is True
+    assert second["idempotent"] is True
+    assert second["status"] == "blocked"
+    # The original block must be preserved untouched.
+    from hermes_cli import kanban_db as kb
+    conn = kb.connect()
+    try:
+        assert kb.get_task(conn, worker_env).status == "blocked"
+    finally:
+        conn.close()
+
+
+def test_block_dependency_kind_double_block_is_idempotent(worker_env):
+    """kind='dependency' routes to todo (not blocked); a second dependency
+    block must also be treated as an idempotent no-op."""
+    from tools import kanban_tools as kt
+    first = json.loads(
+        kt._handle_block({"reason": "waiting on parent", "kind": "dependency"})
+    )
+    assert first["ok"] is True
+    assert first["status"] == "todo"
+    second = json.loads(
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Correct idempotency hardening for `kanban_block`. Prevents spurious worker failure on double delivery / judge-driven retry by distinguishing unknown vs already-blocked vs already-finished.

**What it does**
- `tools/kanban_tools.py:91` after `block_task` returns `False`, re-fetches `kb.get_task(conn, tid)` instead of returning ambiguous `unknown id or not in running/ready`. If `landed is None` → `unknown id`; if `landed.status in ("blocked","triage")` or `todo+block_kind=="dependency"` → returns `_ok(... idempotent=True, status=..., block_kind=..., note="... existing block preserved")` preserving original block; else returns distinct `status '...' , not running/ready (already finished cannot be blocked)`.
- Adds four tests in `tests/tools/test_kanban_tools.py:9`: `already_blocked_is_idempotent`, `dependency_kind_double_block_is_idempotent`, `unknown_id_gets_distinct_error`, `finished_task_gets_distinct_error` — verifies preservation and distinct messages.

**Non-blocking notes**
- `blocked_family` definition `(blocked,triage) or (todo+dependency)` mirrors `block_task` routing — if that routing later adds new blocked-family states, this disambiguation must stay in sync; consider centralizing predicate.
- Idempotent success returns `run_id` from `kb.latest_run` — confirm `latest_run` after an idempotent no-op still points to the original block's run (not a new run); tests assert status but not `run_id` stability.
- Error for finished task includes `landed.status` repr — helpful distinctness, but caller should not rely on parsing repr for branching.

PR #98182, opened 2026-08-29
fix(context): discover Hermes project files in subdirectories

AlekTkT:
## Summary

- discover `.hermes.md` and `HERMES.md` when tools enter a subdirectory
- align progressive hint discovery with the filenames supported by the startup prompt builder
- seed their digests through the existing shared filename list, preserving deduplication

## Reproduction

A session starts above a project, then uses `read_file` inside a child directory that contains only `.hermes.md`.

```text
origin/main: result is None, canary loaded = false
PR branch:   result exists, canary loaded = true
```

This is a real setup on the reporting host: the agent starts in `/home/bernard`, while a child project uses `.hermes.md` as its highest-priority Hermes context. Before this patch, progressive discovery ignored that file and could load secondary `AGENTS.md` / `CLAUDE.md` guidance instead.

## Safety

The new files use the existing path for:

- context threat scanning
- 8,000-character truncation
- digest deduplication
- excluded-directory checks
- one-time directory loading

No new loader or precedence mechanism is introduced.

## Test plan

- [x] RED observed for `.hermes.md`
- [x] RED observed for `HERMES.md`
- [x] two targeted tests pass
- [x] `pytest tests/agent/test_subdirectory_hints.py tests/agent/test_subdirectory_hints_tilde.py` — 30 passed
- [x] standalone real-module canary changes from absent to loaded
- [x] `ruff check agent/subdirectory_hints.py tests/agent/test_subdirectory_hints.py`
- [x] `git diff --check`

### DIFF
--- a/agent/subdirectory_hints.py
+++ b/agent/subdirectory_hints.py
@@ -28,6 +28,7 @@ logger = logging.getLogger(__name__)
 # Same filenames as prompt_builder.py but we load ALL found (not first-wins)
 # since different subdirectories may use different conventions.
 _HINT_FILENAMES = [
+    ".hermes.md", "HERMES.md",
     "AGENTS.override.md",
     "AGENTS.md", "agents.md",
     "CLAUDE.md", "claude.md",
--- a/tests/agent/test_subdirectory_hints.py
+++ b/tests/agent/test_subdirectory_hints.py
@@ -53,6 +53,21 @@ class TestSubdirectoryHintTracker:
         assert result is not None
         assert "Frontend rules" in result
 
+    @pytest.mark.parametrize("filename", [".hermes.md", "HERMES.md"])
+    def test_discovers_hermes_md(self, tmp_path, filename):
+        """Hermes project context must follow tools into subdirectories."""
+        project = tmp_path / "service"
+        project.mkdir()
+        (project / filename).write_text("Hermes project rules")
+
+        tracker = SubdirectoryHintTracker(working_dir=str(tmp_path))
+        result = tracker.check_tool_call(
+            "read_file", {"path": str(project / "main.py")}
+        )
+
+        assert result is not None
+        assert "Hermes project rules" in result
+
     def test_no_duplicate_loading(self, project):
         """Same directory should not be loaded twice."""
         tracker = SubdirectoryHintTracker(working_dir=str(project))


liyangbing:
<h3>Review</h3><p>This closes a real context-discovery gap: subdirectory hint tracking now considers the same Hermes-specific filenames supported by the startup prompt path, while reusing the existing scanning, truncation, digest de-duplication, excluded-directory, and one-time-load mechanisms. The two canary tests directly cover the previously missing files.</p><h3>Suggested contract</h3><ul><li>Define precedence and merge order explicitly. Adding <code>.hermes.md</code> and <code>HERMES.md</code> before the other names changes which guidance is presented first when several files exist in one directory; test the intended ordering and whether all files are concatenated or only prioritized.</li><li>Test case-insensitive filesystems. On Windows, <code>.hermes.md</code> and <code>HERMES.md</code> may resolve to the same file; verify digest/path de-duplication prevents duplicate context and that case variants remain deterministic on macOS/Linux.</li><li>Exercise nested traversal boundaries: starting above the project, entering several child directories, changing back to a parent, repeated tool calls, symlinked directories/files, and paths outside the working root. Ensure hints cannot escape the intended root or bypass excluded-directory checks.</li><li>Verify the existing safety limits for the new filenames: threat scanning, 8,000-character truncation, decode/permission errors, binary content, empty files, and a file that changes between discovery and read.</li><li>Cover interactions with <code>AGENTS.override.md</code>, <code>AGENTS.md</code>, <code>CLAUDE.md</code>, and lowercase variants, including conflicting instructions and duplicate content across ancestor directories.</li><li>Keep loaded-state invalidation explicit for long-running sessions. If a project context file is edited or replaced after its directory was marked loaded, define whether the tracker refreshes, reports the change, or intentionally keeps the first snapshot.</li></ul><h3>Regression matrix</h3><ul><li><code>.hermes.md</code>, <code>HERMES.md</code>, both together, case variants, empty/malformed/binary files, and duplicate content</li><li>ancestor/child/sibling directory discovery, repeated calls, tool path changes, symlink/traversal, and root confinement</li><li>all supported hint filenames with precedence/conflicts and deterministic concatenation order</li><li>threat scan, size/truncation, encoding/permission, read race, excluded directory, and digest de-duplication</li><li>Windows/macOS/Linux filesystem behavior plus edit-after-load and long-running session refresh</li></ul><p>The reported 30 focused passes and standalone canary are a good baseline. Precedence, case-insensitive duplicate paths, and nested/root-boundary tests would make the new context source safer to rely on for every tool call.</p>

PR #98188, opened 2026-08-29
fix(ci): verify gh-image extension binary

unsupportedpastels:
## What does this PR do?

Pins the executable content used by the privileged E2E-evidence publisher.

`gh-image` is a binary GitHub CLI extension, so `gh extension install --pin` accepts its release tag rather than a commit SHA. The workflow still requests `v1.2.0`, then verifies the installed Linux amd64 binary against the release asset's immutable SHA-256 before any `gh image` command can run.

## Why

The publisher runs through `workflow_run` with pull-request write access and later receives `GH_IMAGE_SESSION_TOKEN`. A mutable or replaced release asset should not execute in that job merely because its tag still says `v1.2.0`.

The pinned provenance is:

- tag `v1.2.0` → commit `44f4b93ecbbe22de6c45fa2f62f519aee564ca8c`
- `linux-amd64` SHA-256 → `0505f8c46d63bd603a445fdbfdd6be45e75a80778d97f1edf3580697fa6b7919`

This follows the repository's supply-chain pinning posture without changing the publisher's permissions or behavior.

## Related work

- #69699 introduced the inline evidence publisher.
- #69793 authenticated extension installation while retaining the release tag.
- No existing PR was found that verifies the installed extension content.

## Type of change

- [x] Security hardening
- [x] CI configuration
- [ ] Runtime behavior change

## How to test

- [x] Parsed `.github/workflows/publish-e2e-evidence.yml` with PyYAML.
- [x] Installed `drogers0/gh-image@v1.2.0` under an isolated `XDG_DATA_HOME` and verified the exact workflow path/hash command reports `gh-image: OK`.
- [x] `scripts/run_tests.sh tests/ci/test_publish_e2e_evidence.py tests/ci/test_classify_changes.py -q` — 65 passed.
- [x] `git diff --check`.

## Scope

This deliberately does not change evidence validation, upload behavior, tokens, or permissions. It only rejects unexpected `gh-image` executable bytes before use.

### DIFF
--- a/.github/workflows/publish-e2e-evidence.yml
+++ b/.github/workflows/publish-e2e-evidence.yml
@@ -32,10 +32,17 @@ jobs:
           persist-credentials: false
 
       # v1.2.0 resolves to 44f4b93ecbbe22de6c45fa2f62f519aee564ca8c.
+      # Binary extensions require the release tag for gh --pin, so enforce the
+      # immutable linux-amd64 binary digest before any gh-image use.
       - name: Install gh-image
         env:
           GH_TOKEN: ${{ github.token }}
-        run: gh extension install drogers0/gh-image --pin v1.2.0
+        run: |
+          set -euo pipefail
+          gh extension install drogers0/gh-image --pin v1.2.0
+          GH_IMAGE_BIN="${XDG_DATA_HOME:-$HOME/.local/share}/gh/extensions/gh-image/gh-image"
+          echo "0505f8c46d63bd603a445fdbfdd6be45e75a80778d97f1edf3580697fa6b7919  $GH_IMAGE_BIN" \
+            | sha256sum --check -
 
       - name: Download and attach evidence
         env:


liyangbing:
<h3>Review</h3><p>This strengthens a privileged <code>workflow_run</code> publisher by checking the installed Linux amd64 <code>gh-image</code> bytes before the first use. The explicit tag→commit and digest provenance, <code>set -euo pipefail</code>, and fail-closed hash check are useful supply-chain evidence.</p><h3>Suggested contract</h3><ul><li>Distinguish install-time trust from use-time trust. The workflow installs from a mutable release tag and verifies the binary afterward; confirm that <code>gh extension install</code> cannot execute repository-provided build/install hooks or source during installation. If it can, verify the release asset before installation or use a trusted prebuilt artifact with immutable provenance.</li><li>Pin and verify the exact artifact identity, not only a digest comment: record the repository, tag/commit, asset name, runner OS/architecture, download URL, and expected hash in a reviewable source of truth. Fail if the runner is not Linux amd64 rather than hashing a path with that assumption.</li><li>Make the binary path robust to <code>GH_CONFIG_DIR</code>, <code>XDG_DATA_HOME</code>, gh version/layout changes, symlinks, and spaces. Assert the resolved file is a regular executable owned by the expected extension directory before hashing and execution.</li><li>Test tampering and absence: altered bytes, missing file, wrong architecture, unexpected extra executable, hash-tool failure, extension install failure, and a hash mismatch must stop the job before any <code>gh image</code> command or evidence upload.</li><li>Keep secrets and permissions bounded after the check. Verify the extension cannot receive <code>GH_IMAGE_SESSION_TOKEN</code> or unrelated credentials during installation, and that a failed verification leaves no usable partially installed extension in a later step/cache.</li></ul><h3>Regression matrix</h3><ul><li>expected Linux amd64 asset, non-amd64 runner, missing/wrong-path binary, symlink, and extra executable</li><li>exact digest, one-byte mutation, truncated/empty binary, hash utility failure, and install failure</li><li>mutable tag/changed release asset, commit/asset mismatch, and provenance metadata mismatch</li><li>install-time hook/source execution versus post-install first-use verification</li><li>failure at every gate with no <code>gh image</code> invocation, upload, or secret exposure</li></ul><p>The reported 65 CI tests, isolated install, and exact hash witness are a good baseline. Proving that verification protects the installation phase—not only the later invocation—is the key remaining supply-chain boundary for this privileged workflow.</p>

PR #98190, opened 2026-08-29
fix(gateway): warn on missing scope only for Relay egress

AlekTkT:
## Summary

- resolve the actual delivery transport before warning about a missing tenant scope
- suppress the Relay-specific warning for native Discord/Slack/etc. adapters
- preserve the warning unchanged when Relay really fronts the logical platform

## Reproduction

A native Discord gateway receives synthetic background-process events for a group chat. The events have no `scope_id`, which is valid for the native adapter, but every event logs:

```text
scoped relay egress may be declined by the connector's tenant guard
```

Measured on the reporting host: 40 warnings in under five hours, while Discord is native and Relay is not configured. All corresponding loop replies delivered successfully.

## Before / after

```text
              origin/main   PR branch
native Discord      1 warning    0 warnings
Relay-fronted       1 warning    1 warning
```

The patch reuses `resolve_delivery_transport`, whose contract is native-first and Relay-only-when-fronted. It does not change source reconstruction, routing, authorization, provenance, or the fail-closed Relay diagnostic.

## Test plan

- [x] RED observed: native transport still emitted the Relay warning
- [x] Relay control remained green before the patch
- [x] two focused transport-contract tests pass
- [x] full `test_relay_delivery_followups.py` — 29 passed
- [x] delivery resolver + followups — 42 passed
- [x] standalone real-module before/after preserves the Relay warning
- [x] ruff clean
- [x] `git diff --check`

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -26098,19 +26098,22 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
 
         scope_id = str(evt.get("scope_id") or "").strip() or None
         if scope_id is None and chat_type not in ("dm", "thread"):
-            # Reconstructed (non-persisted) source for a scoped chat with no
-            # scope discriminator: on a relay-fronted deployment the
-            # connector's fail-closed tenant guard may decline the reply
-            # unless user_id resolves it (resolveByUser). Don't fail here —
-            # DMs and author-bound scoped chats still route, and native
-            # adapters don't need scope_id — but say so, so a post-restart
-            # egress decline isn't silent.
-            logger.warning(
-                "Synthetic event source for %s chat=%s (%s) reconstructed "
-                "without scope_id; scoped relay egress may be declined by "
-                "the connector's tenant guard (user_id fallback only).",
-                platform_name, chat_id, chat_type,
+            # Only Relay egress needs a tenant scope. Native adapters route on
+            # their own authenticated platform connection, so warning for them
+            # is both noisy and misleading. Reuse the canonical resolver to
+            # preserve its native-first / Relay-fronting contract.
+            transport = resolve_delivery_transport(
+                platform,
+                self.config,
+                self.adapters,
             )
+            if transport is not None and transport.is_relay:
+                logger.warning(
+                    "Synthetic event source for %s chat=%s (%s) reconstructed "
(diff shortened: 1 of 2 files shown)


liyangbing:
<h3>Review</h3><p>This narrows the warning to the condition it actually describes: the canonical delivery resolver is consulted, native-first delivery no longer emits Relay tenant-scope noise, and a genuinely Relay-fronted Discord path keeps the actionable warning. The paired caplog tests directly protect both sides of that contract.</p><h3>Suggested contract</h3><ul><li>Keep this diagnostic-only change separate from delivery authorization. Add an assertion that native routing, Relay-fronted routing, and a missing/unknown resolver result have identical delivery behavior before and after the warning decision.</li><li>Test resolver inputs as they occur in production: platform enum versus string, adapter aliases, native adapter present but unhealthy, Relay present but not fronting the platform, and a resolver exception. A resolver failure must not accidentally suppress a fail-closed Relay warning or break event handling.</li><li>Define what “Relay egress” means when a platform has multiple adapters or a native adapter is selected dynamically. Verify the warning follows the actual selected transport, not merely the existence of a Relay adapter in the registry.</li><li>Cover warning deduplication and log volume for repeated post-restart synthetic events. The fix removes false positives, but a real Relay path can still emit one warning per event; rate limiting or correlation may be needed for operator usability.</li><li>Check log safety and observability: sanitize platform/chat/user identifiers as required, preserve structured fields for diagnosis, and ensure the warning remains visible at the intended logger/level across native, Relay, and fallback source reconstruction.</li></ul><h3>Regression matrix</h3><ul><li>native Discord/Slack/other adapters, Relay-fronted platform, Relay not configured, and unknown platform</li><li>scope present, scope absent, DM/thread/group, source reconstruction, and user-id fallback</li><li>enum/string/alias resolver inputs, multiple adapters, unhealthy native adapter, and resolver exception</li><li>successful delivery, fail-closed tenant decline, repeated event bursts, restart/reconnect, and warning volume</li><li>structured log fields, identifier redaction, level/filter behavior, and no change to authorization or routing</li></ul><p>The reported 29 follow-up passes, 42 resolver/follow-up passes, and native/Relay before-after controls are a useful baseline. Resolver-failure and selected-transport cases would ensure the quieter native path does not weaken the real Relay diagnostic.</p>

PR #80728, opened 2026-08-07
feat(plugins): security scanning for plugin install/update — inspired by Claude Cowork

teknium1:
## Summary

`hermes plugins install` and `hermes plugins update` now security-scan the plugin tree before it can run — safe installs pass silently, suspicious ones require explicit confirmation, and malicious ones are blocked outright.

Inspired by Claude Cowork's skill & plugin security scanning ([release notes, Aug 6 2026](https://support.claude.com/en/articles/12138966-release-notes), [feature doc](https://support.claude.com/en/articles/15927065)): Cowork automatically checks third-party skills and plugins for malicious content on upload/edit and returns pass / warn / fail, with fail = blocked and not admin-overridable.

Hermes already scans hub-installed **skills** via `tools/skills_guard.py`, but `hermes plugins install` cloned and activated arbitrary Git repos completely unscanned — and plugins are the *more* dangerous surface: they run Python in-process with the agent, with full access to its environment.

## How Cowork does it → how Hermes adapts it

| Cowork | Hermes |
|---|---|
| Scan on upload/edit, background, cached | Scan the temp clone before it's moved into `~/.hermes/plugins/`; re-scan on `hermes plugins update` |
| **Pass** → installs silently | `safe` → installs normally |
| **Warn** → caution banner, user acknowledges | `caution` → findings report + `Install anyway? [y/N]` (or `--force`; dashboard gets structured findings) |
| **Fail** → blocked, no override | `dangerous` → blocked; `--force` does **not** override (same policy as skills_guard) |
| LLM-based review in isolated env | Static analysis reusing the existing `skills_guard` threat-pattern engine — local, free, deterministic |
| Off by default (Enterprise setting) | **On by default**, `plugins.scan_on_install: false` opts out |

## Changes

- `tools/plugin_guard.py` (new): plugin-adapted scanner over the `skills_guard` pattern engine. Key adaptation: plugins are *expected* to read their own API keys from env (`requires_env` pattern) and call provider APIs with them — those pattern ids are exempt on code files but still enforced on docs/config files, where they're an injection/social-engineering signal. Kept at full strength: foreign credential-store access (`~/.ssh`, `~/.aws`, `~/.hermes/.env` reads), reverse shells, destructive commands, persistence, obfuscated execution, symlink escapes. Plugin-sized structural limits (400 files / 10MB); `.git`, `__pycache__`, `node_modules`, venvs excluded.
[...]

PR #98191, opened 2026-08-29
fix(plugins): scan dashboard plugin updates

unsupportedpastels:
## What does this PR do?

Makes dashboard plugin updates use the same post-update scan and disable policy as `hermes plugins update`.

The shared `_scan_updated_plugin()` helper now owns the existing behavior:

- honor the plugin-scan config gate;
- report suspicious or dangerous updated trees;
- automatically disable a plugin when the updated revision is classified as dangerous;
- leave clean updates and scanner-disabled installations unchanged.

Both CLI and dashboard update paths call that helper after a successful pull.

## Why

The plugin scanner added in #80728 covered installation and CLI updates, but `dashboard_update_user_plugin()` independently pulled executable plugin code without invoking it. Updating the same plugin through two supported surfaces therefore produced different safety/UX behavior.

This is correctness and surface parity for an advisory heuristic, not a claim that the scanner is a security boundary.

## Related work

- #80728 introduced plugin install/update scanning and the CLI policy.
- #84975 hardened the shared plugin pull/autostash lifecycle.
- No existing PR was found that applies the established policy to dashboard updates.

## Type of change

- [x] Bug fix
- [x] Tests
- [ ] New feature

## How to test

The regression creates a temporary unpinned Git plugin, advances its remote to a harmless scanner fixture representing a dangerous revision, updates it through the dashboard helper, and verifies that:

- the update completes and returns a JSON-serializable dashboard response;
- the new revision is installed;
- the plugin is added to the disabled set.

Test-first evidence: the focused regression failed on the original dashboard path because no disabled plugin config was written, then passed after both update surfaces shared the helper.

- [x] `scripts/run_tests.sh tests/hermes_cli/test_plugin_install_ref.py tests/hermes_cli/test_plugins_cmd.py tests/tools/test_plugin_guard.py -q` — 86 passed.
- [x] `git diff --check`.

## Scope

Pinned-plugin refusal, git pull/autostash, install metadata, bytecode cleanup, example copying, and dashboard response shape are unchanged.

### DIFF
--- a/hermes_cli/plugins_cmd.py
+++ b/hermes_cli/plugins_cmd.py
@@ -139,6 +139,48 @@ def _scan_plugin_tree(plugin_dir: Path, identifier: str, *, force: bool, scan_de
     return result
 
 
+def _scan_updated_plugin(plugin_dir: Path, identifier: str, console) -> None:
+    """Re-scan an updated plugin and disable dangerous revisions.
+
+    Updates have already mutated the checkout, so caution and dangerous
+    findings are reported rather than blocking the update. A dangerous result
+    disables the plugin, matching the CLI update policy for every update
+    surface.
+    """
+    if not _scan_on_install_enabled():
+        return
+
+    from tools.plugin_guard import (
+        format_scan_report,
+        scan_plugin,
+        should_allow_plugin_install,
+    )
+
+    scan_result = scan_plugin(plugin_dir, source=identifier)
+    allowed, reason = should_allow_plugin_install(scan_result)
+    if allowed is True:
+        return
+
+    console.print()
+    console.print(
+        f"[yellow]⚠ Security scan flagged the updated plugin:[/yellow] {reason}",
+    )
+    console.print(format_scan_report(scan_result))
+    if scan_result.verdict == "dangerous":
+        enabled = _get_enabled_set()
+        disabled = _get_disabled_set()
+        if identifier in enabled or identifier not in disabled:
+            enabled.discard(identifier)
+            disabled.add(identifier)
+            _save_enabled_set(enabled)
+            _save_disabled_set(disabled)
+        console.print(
+            f"[red]Plugin '{identifier}' has been disabled.[/red] Review the "
+            f"findings, then re-enable with `hermes plugins enable {identifier}` "
+            f"if you trust them.",
+        )
+
+
 # Minimum manifest version this installer understands.
 # Plugins may declare ``manifest_version: 1`` in plugin.yaml;
 # future breaking changes to the manifest schema bump this.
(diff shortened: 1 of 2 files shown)


liyangbing:
<h3>Review</h3><p>This removes the behavioral gap between the CLI update path and <code>dashboard_update_user_plugin()</code>: both now scan the post-pull tree and disable a dangerous revision while preserving the distinction between advisory findings and a hard install gate. The regression also verifies the unpinned dashboard update reaches the new revision and writes the disabled configuration.</p><h3>Suggested contract</h3><ul><li>Make the dashboard result truthful. <code>_scan_updated_plugin</code> writes a warning to a new <code>Console()</code>, but the JSON response still reports only a successful update; return a structured scan verdict/reason and <code>disabled</code> state so UI/API callers cannot present a dangerous update as an ordinary success.</li><li>Clarify runtime behavior after a dangerous update. Disabling future discovery does not necessarily unload code already imported or stop a running plugin; verify whether dashboard routes, workers, websockets, and cached plugin instances are removed or explicitly document the residual exposure.</li><li>Keep the pull→scan boundary race-safe: scan the exact checkout/revision that was installed, reject symlink/submodule/worktree escapes, and test a tree changed between pull completion and scan. Record the scanned commit and scanner result with install metadata.</li><li>Test all update outcomes: clean, caution, dangerous, scanner-disabled, scan exception, and config-write failure. A scan failure must not silently look like a clean dashboard update, and a failed disable write must be visible and recoverable.</li><li>Verify parity with the CLI for pinned versus unpinned plugins, autostash/rollback behavior, bytecode cleanup, example-file copying, and update-with-no-changes. The shared helper should not change the dashboard response shape or console/error semantics unexpectedly.</li><li>Ensure findings are redacted and bounded before reaching dashboard logs or API responses; plugin content may contain credentials, URLs, or attacker-controlled strings.</li></ul><h3>Regression matrix</h3><ul><li>CLI and dashboard clean/caution/dangerous updates, pinned/unpinned revisions, and scanner disabled</li><li>dangerous update with loaded routes, background workers, websocket handlers, and subsequent restart/re-enable</li><li>revision changed during scan, symlink/submodule escape, scan exception, and malformed plugin tree</li><li>disabled-set write failure, metadata failure, rollback/autostash, unchanged update, and bytecode cleanup</li><li>structured dashboard response, logs, redaction, finding limits, and exact installed/scanned commit identity</li></ul><p>The reported 86 focused passes and test-first dangerous-update regression are a solid baseline. Returning the scan outcome and defining what happens to already-loaded plugin code are the key remaining operational safety boundaries for a dashboard update.</p>

PR #85057, opened 2026-08-13
fix(plugins): sanitize nested pack config (#85050)

michaelversluis:
Fixes #85050

## Problem

Plugin-pack config validation and export sanitization only inspected top-level keys. Nested dictionaries, lists, and tuple-backed YAML values could therefore carry secrets, capability-consent state, or deprecated `allow_*` trust gates. Recursive YAML aliases could also crash sanitization with `RecursionError`.

## Solution

- Apply one recursive config-key policy to import validation and export sanitization.
- Reject forbidden keys at any depth during import/seed validation.
- Strip forbidden keys at any depth during export.
- Traverse tuple-backed `!!pairs` / `!!omap` values and emit YAML-safe lists.
- Detect cyclic aliases per traversal: imports fail with `PackError`; exports omit the cyclic branch without leaking data.
- Preserve normal nested structures and shared acyclic values.

## Tests

- Verified RED on unpatched behavior for nested dict/list import and export cases.
- Verified RED for `!!pairs` / `!!omap` and tuple export cases.
- Verified RED for cyclic dict/list aliases (`RecursionError`).
- `scripts/run_tests.sh tests/hermes_cli/test_plugin_packs.py -q` — 48 passed.
- Ruff on both changed files — passed.
- `git diff --check` — passed.
- Independent fail-closed review — passed after tuple and cyclic-alias follow-ups.

## Scope

Only `hermes_cli/plugin_packs.py` and `tests/hermes_cli/test_plugin_packs.py` are changed.

_NL: deze fix zorgt dat secrets en capability-toestemming ook diep genest nooit via plugin packs kunnen meereizen._

PR #98195, opened 2026-08-29
fix(plugins): validate plugin pack redirects

unsupportedpastels:
## What does this PR do?

Makes remote plugin-pack loading validate the initial URL and every redirect before sending the next request.

The loader now:

- requires HTTPS on the initial request and every redirect hop;
- applies the shared URL-safety policy to reject loopback, private, link-local, and metadata destinations;
- uses the shared connect-time guarded `httpx` client;
- resolves and validates redirect `Location` values before `httpx` follows them;
- caps redirects at five;
- preserves valid public HTTPS absolute and relative redirects;
- keeps the existing timeout, 1 MiB payload limit, parsing, exact plugin-SHA requirement, and local-file behavior.

## Why

Plugin packs intentionally accept operator-supplied HTTPS URLs. The original loader used `follow_redirects=True` directly, so the initial scheme requirement did not apply to later destinations. This is robustness and network-boundary hardening for an explicit operator action, not an unauthenticated vulnerability claim.

## Related work

- #84988 introduced plugin packs and their supply-chain checks.
- #85057 sanitizes nested pack config but retains the original redirect behavior.
- Existing Skills Hub downloads already use the shared SSRF-safe client and per-hop URL policy; this change follows that established pattern.

## Type of change

- [x] Security hardening
- [x] Robustness fix
- [x] Tests

## How to test

Test-first evidence: the new redirect contract failed on the original loader because it followed redirects to loopback/private/link-local and HTTP destinations. It passed after applying the guarded client and per-hop validation.

The tests verify:

- loopback, RFC1918, and link-local redirects are rejected before a second request;
- HTTPS-to-HTTP downgrade is rejected;
- relative and absolute public HTTPS redirects still load a valid pack;
- requested URL order proves blocked targets are never contacted.

- [x] `scripts/run_tests.sh tests/hermes_cli/test_plugin_packs.py tests/tools/test_url_safety.py tests/hermes_cli/test_plugin_index_search.py -q` — 141 passed.
- [x] `git diff --check`.

## Scope

This does not change plugin installation, capability consent, pack configuration, plugin revision pinning, proxy configuration, or local pack files.

### DIFF
--- a/hermes_cli/plugin_packs.py
+++ b/hermes_cli/plugin_packs.py
@@ -45,6 +45,7 @@ import sys
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, List, Optional
+from urllib.parse import urlparse
 
 logger = __import__("logging").getLogger(__name__)
 
@@ -62,6 +63,7 @@ _SECRET_KEY_RE = re.compile(
 _RESERVED_ENTRY_KEYS = frozenset({"granted_capabilities", "capabilities_consent"})
 
 _MAX_PACK_BYTES = 1 * 1024 * 1024  # a pack is a small manifest, not a payload
+_MAX_PACK_REDIRECTS = 5
 _FETCH_TIMEOUT = 15.0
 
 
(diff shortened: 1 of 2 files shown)


liyangbing:
<h3>Review</h3><p>This is a meaningful network-boundary hardening: the initial pack URL and each redirect target are checked for HTTPS and the shared SSRF policy before the next request, redirects are bounded, and transport errors no longer echo a caller-controlled URL. The tests also assert request order, so blocked private/HTTP targets are not contacted.</p><h3>Suggested contract</h3><ul><li>Test the full URL lifecycle, not only the redirect destination: uppercase/mixed-case schemes, whitespace, userinfo, ports, fragments, encoded host forms, IPv4/IPv6 literals, DNS rebinding, and lookalike public domains should have explicit outcomes.</li><li>Verify relative redirect resolution against the original URL's path, scheme, and port, including <code>Location</code> values with query-only/path-relative forms and malformed locations. Confirm every resolved hop is validated before transport follow.</li><li>Keep provenance safe after a successful fetch. <code>parse_pack(..., source=path_or_url)</code> may retain a caller-supplied URL containing userinfo or sensitive query data; ensure review errors, logs, parsed metadata, and install records use a redacted source.</li><li>Define failure semantics for redirect-hook exceptions, too many redirects, missing <code>Location</code>, invalid JSON/YAML, oversized compressed responses, streaming bodies, timeouts, and a blocked later hop. No partial pack should reach parsing or installation.</li><li>Test the shared SSRF client configuration itself: DNS resolution per hop, IPv4/IPv6, proxy/environment settings, connection versus read timeout, and whether a redirect response can be consumed or retried after the guard rejects it.</li><li>Preserve local-file behavior and exact plugin-SHA/revision checks, while proving an HTTPS pack with a valid redirect cannot bypass capability/installation policy through metadata or a redirected content type.</li></ul><h3>Regression matrix</h3><ul><li>direct public HTTPS, relative/absolute public HTTPS redirects, HTTP downgrade, loopback, RFC1918, link-local, metadata, IPv6, and DNS-rebinding cases</li><li>multi-hop chains, five-hop limit, loop/cyclic cursor, malformed/missing location, query/path-relative redirects, and port changes</li><li>userinfo/query-bearing URLs, redacted errors/logs/provenance, invalid/oversized/truncated payloads, and timeout failures</li><li>request-order assertions proving no blocked target receives a connection, including later-hop failures</li><li>local files, SHA-pinned packs, capability consent, and successful install after a safe redirect</li></ul><p>The reported 141 focused passes and test-first redirect evidence are strong. Redacted provenance and malformed/multi-hop/DNS edge cases would complete the safety contract around this operator-supplied network input.</p>

PR #98198, opened 2026-08-29
test(a2a): isolate the process-global tool registry in plugin tests

c-pompa:
## Problem

Built-in tool modules (e.g. `tools/web_tools.py`) register at import time, and plugin-scoped registrations linger under other `HERMES_HOME` scope keys. When one pytest process collects such a module before `tests/plugins/test_a2a_plugin.py`, the registry is non-empty and `A2AAdapter._advertised_skills()` takes its dynamic path — Agent Card tests advertise whatever leaked in instead of the configured static capabilities.

## Repro on main

`pytest tests/tools/test_web_tools_truncate.py tests/plugins/test_a2a_plugin.py` → `test_path_routed_agent_card_uses_prefix_and_canonical_path` fails: card skills are `{'web'}` instead of `{'research', 'web'}`. (Passes in isolation — ordering-dependent.)

## Fix

Autouse fixture that snapshots, clears, and restores the process-global registry, so every test in the module starts from an empty registry and prior state is restored on teardown.

## Testing

- Combined run: 113 passed (was 1 failed) · standalone: 106 passed


liyangbing:
<h3>Review</h3><p>This isolates the order-dependent A2A test failure at the source of the contamination: the fixture snapshots the process-global registry under its lock, clears both global and scoped maps before each test, and restores them in teardown. That directly protects the static Agent Card capability path from tools imported by an earlier test module.</p><h3>Suggested contract</h3><ul><li>Keep the snapshot/clear/restore operation atomic with respect to registry readers and writers, and add a failure-path test proving restoration runs when the test body raises or fixture setup fails.</li><li>Verify that <code>_tools</code> and <code>_scoped_tools</code> are the complete mutable registry state. If the registry has generation counters, caches, aliases, listeners, or lazy-import bookkeeping, include those or document why they are intentionally excluded.</li><li>Exercise dynamic and scoped lookups while the registry is empty, then verify the saved registrations are visible after teardown to a following test and to another thread/process that shares the test runtime.</li><li>Define interaction with pytest-xdist and module-level import registration. A process-local fixture cannot isolate workers from external state, and clearing registrations after another test has begun can still create a race if tests run concurrently in one process.</li><li>Prefer a public reset/snapshot API or a test-only registry context if available; relying on private <code>_lock</code>, <code>_tools</code>, and <code>_scoped_tools</code> makes this fixture tightly coupled to implementation details.</li><li>Cover the actual A2A decision branches: empty registry/static capabilities, non-empty dynamic capabilities, scoped tool registrations, and restoration after a test that imports built-in or plugin tools.</li></ul><h3>Regression matrix</h3><ul><li>web-tools imported before A2A versus A2A collected first, in the combined reproduction order</li><li>empty/global/scoped registry, dynamic Agent Card path, static configured capabilities, and missing capabilities</li><li>test pass, assertion failure, fixture exception, and teardown restoration</li><li>parallel threads, xdist workers, repeated module collection, and plugin import side effects</li><li>registry state before/after each test, including aliases/caches/listeners if present</li></ul><p>The reported combined 113-pass result and isolated 106-pass result are useful evidence. Restoration-on-failure and concurrent/parallel collection coverage would ensure this test isolation remains reliable as the registry evolves.</p>

c-pompa:
Thanks — good contract list. Two points adopted on the branch (rebased on current main), two pushed back on:

**Adopted:**
- **Failure-path test**: `test_registry_fixture_restores_state_when_test_body_raises` drives the fixture generator the way pytest does, registers a sentinel tool, raises mid-body, and asserts the teardown still restored the snapshot.
- **Completeness of registry state**: documented in the fixture docstring — only `_tools`/`_scoped_tools` are snapshot/cleared; `_toolset_aliases`, `_toolset_checks`, and `_plugin_override_policy` are configuration/ownership bookkeeping, not tool leakage, and are deliberately excluded (they're not what leaks between imported tool modules).

**Disagree (with reasons):**
- **Public reset API**: I'd keep the private-attribute coupling. This is a test fixture, and coupling a fixture to registry internals is the desired kind of coupling — a public `reset()` on a production singleton invites production code to use it. The docstring now states the intent explicitly.
- **pytest-xdist**: the registry is a process-local singleton per worker, so cross-worker contamination can't occur — there's no shared state for the fixture to guard. Also documented in the docstring rather than coded.

c-pompa:
Follow-up push while completing the triage items:

- 7d30372257: `test_card_skills_reflect_registry_state_with_fallback_and_filter` — covers both `_advertised_skills()` decision paths that read the process-global tool registry: the dynamic path (populated registry → toolsets with tool-name tags, honoring the `advertised_toolsets` filter) and the empty-registry fallback to the configured static list. Together with the fixture failure-path test, both sides of the isolation contract are now pinned.

PR #98199, opened 2026-08-29
test(video-gen): stub the lazy _fal_client global directly in fal tests

c-pompa:
## Problem

The `with_fake_fal` fixtures stub `sys.modules['fal_client']` and reset `_fal_client = None` so the next `_load_fal_client()` re-imports. But `import_fal_client()` calls `lazy_deps.ensure()` first, which raises `FeatureUnavailable` when the fal-client extra isn't installed (CI, fresh dev envs) — the stubbed import is never reached and 8 tests fail.

## Repro on main

Without fal-client installed: `HERMES_DISABLE_LAZY_INSTALLS=1 pytest tests/plugins/video_gen/test_fal_plugin.py` → 8 failed (TestFamilyRouting ×5, TestUpscalePass ×3).

## Fix

Set the module-global `_fal_client` to the stub via `monkeypatch.setattr` in both fixtures — which also restores the previous value on teardown (the bare `= None` assignment leaked across tests).

## Testing

- No-fal-client env: 44 passed (was 8 failed) · with real fal-client installed: 44 passed

### DIFF
--- a/tests/plugins/video_gen/test_fal_plugin.py
+++ b/tests/plugins/video_gen/test_fal_plugin.py
@@ -161,9 +161,13 @@ class TestFamilyRouting:
         fake.submit = _submit  # type: ignore
         monkeypatch.setitem(sys.modules, "fal_client", fake)
 
-        # Reset the lazy global so it picks up our stub
+        # Point the lazy global straight at our stub. Going through
+        # sys.modules alone is not enough: import_fal_client() calls
+        # lazy_deps.ensure() first, which raises when the fal-client
+        # package is not installed (it is a lazy extra, absent from CI
+        # and fresh dev envs) — the stubbed import is never reached.
         from plugins.video_gen import fal as fal_plugin
-        fal_plugin._fal_client = None
+        monkeypatch.setattr(fal_plugin, "_fal_client", fake)
         # Also reset the managed client cache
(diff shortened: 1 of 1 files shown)


Enough1122:

**Overall:** Minimal test-fixture fix for `fal_client` lazy-install gate. Direct global stubbing is the correct remedy when `lazy_deps.ensure()` would otherwise short-circuit `sys.modules` injection.

**What it does**
- `tests/plugins/video_gen/test_fal_plugin.py:14` / `:25` changes `with_fake_fal` helpers from `fal_plugin._fal_client = None` to `monkeypatch.setattr(fal_plugin, "_fal_client", fake)` — points the lazy global straight at the fake so `import_fal_client()` never calls `lazy_deps.ensure()` → `FeatureUnavailable` path when extra absent (CI, `HERMES_DISABLE_LAZY_INSTALLS=1`).

**Non-blocking notes**
- Both `with_fake_fal` sites now set `_fal_client = fake` without also inserting into `sys.modules['fal_client']`? They still do `setitem(sys.modules, "fal_client", fake)`, so both seams are patched — redundant but safe.
- `monkeypatch.setattr` is correctly paired with `monkeypatch` lifecycle (auto-restore), unlike bare assignment.

No blocking issues.

PR #98201, opened 2026-08-29
fix(providers): honor provider_routing on a custom:openrouter profile

### LINKED ISSUE #98186: [Bug]: provider_routing.only silently dropped on custom:openrouter → provider drift breaks prompt-cache continuity
## Summary

`provider_routing.only` is silently dropped on the `provider: custom:openrouter` path (a named custom provider with `base_url` pointing at `https://openrouter.ai/api/v1`). The provider lock has no effect and OpenRouter price-sorts every request to whichever reseller is cheapest — so requests for the same conversation drift across provider endpoints from turn to turn.

This isn't a security hole (account-level provider restrictions still apply), but it is a real **UX / performance** regression for long-context agent sessions: **provider drift destroys prompt-cache continuity**.

## Why drift hurts (the visible cost)

OpenRouter's prompt cache is per-provider. When a long conversation keeps bouncing between providers:

- Cache prefix hits collapse — steady ~99% collapses to ~3-5% whenever the endpoint changes.
- Every turn re-prefills the full (possibly 100k+ token) context instead of reusing a cached prefix.
- Result: noticeably slower, more expensive turns on a session that *should* be mostly cache reads.

Observed on a real session (268 calls, ~19M input tokens): with the lock broken the drift pattern gave ~3% cache hits on the bounce turns; once the provider was locked the *same* workload sat at 99-100% prefix hits. Same conversation, same model — only the provider pinning changed.

## Root cause

`provider: custom:openrouter` normalizes to the `custom` provider identity, which resolves to `CustomProfile`. `CustomProfile` inherits the base `build_extra_body` (`providers/base.py`, default returns `{}`), so it never writes `body["provider"]`:

- `providers/base.py:148` — default `build_extra_body` returns `{}`
- `agent/chat_completion_helpers.py:836` — profile branch calls `profile.build_extra_body(provider_preferences=...)`
- Only `plugins/model-providers/openrouter/__init__.py:139-141` sets `body["provider"] = prefs`, and that path is skipped for `custom`

So `provider_routing.only` / `ignore` / `order` / `sort` silently no-op, and the lock the user configured does not hold. Same root family as #89430 (OR-only prefs consumed differently per profile) — here it's the *silent drop* variant rather than the Nous hard-400 variant.

## Expectation

`provider_routing` is a **permanent, global** preference — it should be applied at a level that survives regardless of how the model got selected, and (per the user's framing) either an account/global injection or a per-request one is acceptable. What matters is that the option is not **silently ignored** on a valid configuration path (`custom:openrouter`), because that both breaks the promise of the lock and, more importantly, degrades cache continuity and latency on long agent sessions.

## Workaround (confirmed working)

Switch `model.provider` from `custom:openrouter` to the native `openrouter` profile (keep custom timeouts under `providers.openrouter`). Afterward `provider.only` is injected and the lock holds (verified in the OR activity log: single provider, 99-100% prefix cache).

bsdnn:
## What does this PR do?

A custom provider pointed at an OpenRouter base_url (`provider: custom:openrouter`)
resolves to `CustomProfile`, whose `build_extra_body` returns `{}` and never emits
`body["provider"]`. The transport's provider-routing injection only runs on the
legacy no-profile path (`build_kwargs` bypasses those flags when a profile is
present), so `provider_routing.only/ignore/order/sort` was silently dropped —
OpenRouter price-sorts every turn, the endpoint drifts, and prompt-cache prefix
hits collapse on long sessions.

Fix: in the profile path, inject `provider_preferences` into
`extra_body["provider"]` when the resolved `base_url` is OpenRouter and the profile
did not already set one. **Gated on the base_url**, so non-OpenRouter custom
providers never receive the OpenRouter-only field, and the native OpenRouter
profile (which sets it itself) is unaffected.

Credit to @wjmyjay5-beep for the diagnosis in #98186.

## Related Issue

Fixes #98186

## Type of Change

- [√] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `agent/transports/chat_completions.py` — base_url-gated OpenRouter `provider`
  injection in `_build_kwargs_from_profile`.
- `tests/agent/transports/test_chat_completions.py` — inject / no-inject tests.

## How to Test

    pytest tests/agent/transports/test_chat_completions.py -q

59 pass. New tests: custom profile on an OpenRouter base_url injects
`provider`; a custom profile on any other base_url does not.

## Checklist

- [√] I've read the Contributing Guide
- [√] My commit messages follow Conventional Commits
- [√] I searched for existing PRs / issues
- [√] My PR contains only changes related to this fix
- [√ ] I've run `pytest tests/ -q` and all tests pass
- [√] I've added tests for my changes
- [√] I've tested on my platform: Windows 11

### DIFF
--- a/agent/transports/chat_completions.py
+++ b/agent/transports/chat_completions.py
@@ -844,6 +844,23 @@ class ChatCompletionsTransport(ProviderTransport):
         if profile_body:
             extra_body.update(profile_body)
 
+        # OpenRouter provider-routing also applies when a *custom* profile fronts
+        # an OpenRouter base_url (provider: custom:openrouter). The native
+        # OpenRouter profile emits body["provider"] from build_extra_body, but a
+        # generic CustomProfile does not — so provider_routing.only/ignore/order/
+        # sort would be silently dropped, OpenRouter price-sorts every turn, and
+        # prompt-cache prefix hits collapse on long sessions (#98186). Gate on the
+        # resolved base_url so non-OpenRouter custom providers never get the field,
+        # and never override a "provider" the profile already set.
+        if "provider" not in extra_body:
+            _routing_prefs = params.get("provider_preferences")
+            _routing_base_url = params.get("base_url")
+            if _routing_prefs and _routing_base_url:
+                from utils import base_url_host_matches
+
+                if base_url_host_matches(str(_routing_base_url).lower(), "openrouter.ai"):
+                    extra_body["provider"] = _routing_prefs
+
         # Profile's reasoning/thinking extra_body entries
         if extra_body_from_profile:
             extra_body.update(extra_body_from_profile)
--- a/tests/agent/transports/test_chat_completions.py
+++ b/tests/agent/transports/test_chat_completions.py
@@ -241,6 +241,36 @@ class TestChatCompletionsBuildKwargs:
         )
         assert kw["extra_body"]["provider"] == {"only": ["openai"]}
 
+    def test_custom_profile_on_openrouter_base_url_injects_provider_prefs(self, transport):
+        """#98186: a custom profile pointed at an OpenRouter base_url must still
+        emit body["provider"] so provider_routing.only/ignore/order/sort holds
+        (a CustomProfile's build_extra_body returns {} and would otherwise drop
+        the lock, breaking prompt-cache continuity)."""
+        from providers import get_provider_profile
+        profile = get_provider_profile("custom")
+        msgs = [{"role": "user", "content": "Hi"}]
+        kw = transport.build_kwargs(
+            model="some/model", messages=msgs,
+            provider_profile=profile,
+            base_url="https://openrouter.ai/api/v1",
+            provider_preferences={"only": ["deepinfra"]},
+        )
+        assert kw["extra_body"]["provider"] == {"only": ["deepinfra"]}
+
+    def test_custom_profile_off_openrouter_does_not_inject_provider_prefs(self, transport):
+        """The injection is gated on the OpenRouter base_url — a custom provider
+        on any other endpoint must NOT receive the OpenRouter-only field."""
+        from providers import get_provider_profile
+        profile = get_provider_profile("custom")
+        msgs = [{"role": "user", "content": "Hi"}]
+        kw = transport.build_kwargs(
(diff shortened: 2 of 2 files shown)


liyangbing:
<h3>Review</h3><p>This addresses a concrete profile-path gap: a generic custom profile pointed at OpenRouter now receives the provider-routing body field, while a non-OpenRouter endpoint remains untouched. Gating on the resolved host and retaining the native profile path are the right compatibility boundaries for this fix.</p><h3>Suggested contract</h3><ul><li>Test precedence for every source of <code>extra_body</code>. The new injection happens before <code>extra_body_from_profile</code> is merged, so verify that a profile-supplied <code>provider</code> is never overwritten—or change the merge order if that is the intended authority.</li><li>Exercise all provider preference shapes: <code>only</code>, <code>ignore</code>, <code>order</code>, <code>sort</code>, combined fields, empty mappings, and invalid values. Confirm the payload is passed unchanged to OpenRouter and not shared/mutated between requests.</li><li>Test host matching at the security boundary: official <code>openrouter.ai</code> with ports, subdomains, mixed case, trailing dots, paths/query strings, an attacker-controlled lookalike host, and an HTTP/custom proxy URL. Verify only the intended host is eligible.</li><li>Cover base URL resolution through every caller, including profile endpoint overrides, environment/configured endpoints, and a custom profile whose effective URL differs from the raw parameter.</li><li>Verify interaction with reasoning/tool-call extra bodies, retries, streaming/non-streaming requests, and request logging/redaction. Provider preferences should survive the full transport build without being dropped or accidentally exposed.</li></ul><h3>Regression matrix</h3><ul><li>native OpenRouter, custom profile on OpenRouter, and custom profile on unrelated providers</li><li>provider field absent, supplied by profile body, supplied by profile extra body, and supplied by caller preferences</li><li>official host, port/path/query variants, proxy host, lookalike host, and malformed URL</li><li>each routing preference field alone and in combinations, including empty/invalid values</li><li>streaming/tool-call/retry paths and repeated builds proving no cross-request mutation</li></ul><p>The reported 59 focused passes and the two new inject/no-inject tests are a useful baseline. Precedence coverage and hostile/variant host cases would make the OpenRouter-only injection contract safer to maintain.</p>

PR #98207, opened 2026-08-30
feat(dashboard): runtime reload endpoint for plugin API routes

c-pompa:
## What

Adds `POST /api/plugins/<name>/reload` to the dashboard web server: re-imports one plugin's `dashboard/plugin_api.py` from disk and re-mounts its FastAPI routes in the running process — plugin authors iterate on backend routes without a dashboard restart (uvicorn runs without `--reload`; plugin modules are imported once at boot).

## Why

Today, picking up an edit to a plugin's api file means restarting the dashboard — a slow dev loop, and for operators it drops active sessions to deploy a plugin fix.

## How

`_mount_plugin_api_routes()` is refactored into shared helpers so startup and reload enforce identical policy (bundled/user source gate + `plugins.enabled` allow-list; project plugins never imported — GHSA-5qr3-c538-wm9j / GHSA-mcfc-hp25-cjv7 posture unchanged). Three constraints solved:

1. **`sys.modules` eviction** — plugin modules live under a synthetic name and `sys.modules` is otherwise frozen for the process lifetime, so reload evicts `hermes_dashboard_plugin_<name>` before re-importing.
2. **Before-catch-all insertion** — Starlette matches routes in registration order and the SPA catch-all is already registered at runtime, so re-mounted routes are spliced ahead of it (at startup it isn't registered yet; behavior unchanged).
3. **Safe failure** — the new module is imported *before* stale routes are dropped: a broken mid-edit file yields `{"ok": false, "error": ...}` (HTTP 200, never 500) and the old routes keep serving.

Auth needs nothing new: `auth_middleware` already requires the session token on `/api/`, and `_plugin_api_runtime_gate` 404s disabled/unknown plugins.

## Tests

New `TestPluginAPIRuntimeReload` (6 tests): edit pickup (v1→v2), stale-route removal, broken file → structured error with old routes live, unknown/disabled rejection, unauthenticated 401. `test_web_server.py` + `test_project_plugin_rce_bypass.py`: 203 passed. Also verified live against a temp `HERMES_HOME` with a real user plugin (`{"ok": true, "routes": 15, "stale_removed": 15}`, second reload idempotent).


liyangbing:
<h3>Review</h3><p>This adds a useful development primitive while preserving the existing trust boundary: the reload path re-scans the plugin, applies the same source/enablement gate, imports before removing stale routes, and inserts new routes before the SPA catch-all. The tests also verify authentication, disabled/unknown rejection, stale-route removal, and old-route survival after a syntax failure.</p><h3>Suggested contract</h3><ul><li>Make the reload atomic from the serving process's perspective. Importing the new module before removing old routes protects failures, but a failure while removing or mounting routes can still leave a partial route set; add a snapshot/restore path for the route list and module reference.</li><li>Define concurrency behavior for two simultaneous reload requests and an in-flight request to an old route. A lock or serialized swap should prevent one reload from deleting routes installed by another and should keep the route table coherent.</li><li>Test module lifecycle beyond <code>sys.modules</code> eviction: old module references held by route handlers, dependency objects, background tasks, and imported submodules must not create cross-version behavior or unbounded retained modules.</li><li>Keep the authorization and plugin identity checks fail-closed if configuration loading or a forced rescan fails. The current fallback to empty sets is safe for user plugins, but verify bundled-plugin and project-plugin behavior explicitly at the HTTP boundary.</li><li>Constrain plugin-name and API-prefix handling to the same validated identity used by discovery, and test names containing path-like characters or a prefix collision with another plugin.</li></ul><h3>Regression matrix</h3><ul><li>authenticated reload of one route, added route, removed route, and repeated idempotent reload</li><li>syntax/import failure, missing API file, no-router module, failed route mount, and old-route availability after each failure</li><li>unknown, disabled, project, bundled, and user plugins through both direct helper and HTTP middleware paths</li><li>two concurrent reloads plus requests during the import/swap boundary</li><li>plugin-name/prefix collision, forward-reference models, dependency state, and module cleanup after multiple reloads</li></ul><p>The reported 203 passing tests and live <code>15</code>-route reload result are strong evidence. Atomic swap and concurrent-reload coverage would protect the new mutation path from leaving a running dashboard partially reconfigured.</p>

c-pompa:
Thanks for the careful review — the two structural points were real and are now fixed on the branch (da42d9c008, rebased on current main):

**Adopted:**
- **Atomic route swap**: the stale-removal + mount section now runs under a module-level lock, and the route list is snapshotted before the swap — a mount failure rolls the table back to its exact pre-reload contents. Import stays outside the lock (it's the slow part and touches only its own module). Covered by `test_reload_mount_failure_rolls_back_route_table` (route list restored in order + old routes keep serving).
- **Concurrent reloads**: serialized by the same lock, so one reload can no longer delete routes another just installed.
- **Fail-closed boundary test**: added `test_reload_fail_closed_when_config_read_fails` — if the enabled/disabled sets can't be read, user plugins are treated as not enabled (never fail-open), pinned at both the function and HTTP level.

**On the remaining points:** the empty-set fallback is already fail-closed by construction — for user plugins an empty `enabled_set` means "not in plugins.enabled → refuse", and bundled/project paths don't depend on the sets (project plugins are always refused per GHSA-5qr3-c538-wm9j) — but I agree it deserved a pinning test, which the new test provides. On module lifecycle: old route objects deliberately keep referencing the old module's functions (that's what keeps the old routes serving after a failed re-import, per the docstring); we evict from `sys.modules` so the *next* import is fresh, and won't chase handler-held references without a concrete cross-version symptom — happy to revisit if you've seen one. Name/prefix validation reuses the same discovered plugin identity (the reload looks the plugin up by name from a fresh rescan, so an unknown or path-like name simply never matches); I'd rather not add a second validator that could drift from discovery.

c-pompa:
Follow-up pushes while completing the triage items:

- b244addb90: `test_reload_prefix_boundary_spares_sibling_prefix_plugins` — pins that stale-route removal matches the exact `/api/plugins/<name>` prefix, so a sibling plugin whose prefix merely extends it (`example2` vs `example`) survives a reload untouched. Ninth test in `TestPluginAPIRuntimeReload`, all passing.
- 7d2e21ae02: dropped a duplicated `prefix` assignment left over from the lock-section rewrite (outside vs inside the lock).

PR #98208, opened 2026-08-30
fix(desktop): align "Move to project" submenu with sidebar view filters

ferox-fedm:
The "Move to project" submenu in the session kebab menu offered projects that are not visible in the sidebar project list: auto-projects the user dismissed and projects excluded by the persisted sidebar project filter. Both surfaces read the same $projectTree atom, but only the sidebar applied the view filters (filterVisibleProjects + $sidebarProjectFilter).

MoveToProjectItems now applies the same two view filters before its move-specific constraints (current-project exclusion, !isNoProject, projectRootCwd), so the submenu always offers exactly the project set the sidebar shows.

Tests: new cases in session-actions-menu.test.tsx drive the real Radix menu with the real layout store — a dismissed auto-project and a sidebar-filtered project disappear from the submenu and reappear when re-shown/cleared.

Verification: tsc --noEmit clean; vitest 8/8 in the menu test file; full npm run check (typecheck + eslint + UI suites).

### DIFF
--- a/apps/desktop/src/app/chat/sidebar/session-actions-menu.test.tsx
+++ b/apps/desktop/src/app/chat/sidebar/session-actions-menu.test.tsx
@@ -1,6 +1,9 @@
-import { cleanup, fireEvent, render, screen, waitFor, within } from '@testing-library/react'
+import { act, cleanup, fireEvent, render, screen, waitFor, within } from '@testing-library/react'
 import { atom } from 'nanostores'
 import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import { $dismissedAutoProjectIds, $sidebarProjectFilter } from '@/store/layout'
+import { $projectTree } from '@/store/projects'
 
 import { SessionActionsMenu, SessionContextMenu } from './session-actions-menu'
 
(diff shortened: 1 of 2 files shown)


liyangbing:
<h3>Review</h3><p>This aligns the move destination menu with the sidebar's visible project set by applying <code>filterVisibleProjects</code> and the persisted project filter before the existing move-specific exclusions. The tests drive the real Radix interaction and real layout atoms, including dismiss/re-show and filter/clear transitions, which directly covers the reported stale-menu behavior.</p><h3>Suggested contract</h3><ul><li>Keep this as a view-consistency rule, not an authorization rule: hiding a project must change discoverability in the menu, while <code>moveSessionToProject</code> should still validate the destination against current project state at execution time.</li><li>Add coverage for the existing menu-only constraints after the new filters: current project, no-project/Home, project without a root folder, and a project whose root is provided only by a repository.</li><li>Test filter semantics with an unknown/stale project id, a removed project, and duplicate ids; verify that clearing or changing the project tree cannot leave a previously visible menu item actionable.</li><li>Verify the menu updates while open when a project is dismissed, filtered, deleted, or reloaded, and define whether an active submenu closes or shows the empty-state item when its last target disappears.</li><li>Exercise keyboard navigation and accessibility labels after the target list changes; Radix focus should not remain on a removed menu item or move to the wrong project.</li></ul><h3>Regression matrix</h3><ul><li>visible, dismissed auto, persisted-filtered, current, no-project, and rootless projects</li><li>root path on the project node versus root path on the first valid repository</li><li>empty/unknown/duplicate filter ids and project-tree removal while the submenu is open</li><li>dismiss/re-show, filter/clear, keyboard navigation, and empty-state rendering</li><li>execution-time destination validation after the view state changes</li></ul><p>The 8 focused tests, typecheck, and full check reported by the PR are a good baseline. Testing the existing move constraints and an open-menu deletion would ensure the new view filters do not introduce an actionable stale target.</p>

PR #98210, opened 2026-08-30
fix(kanban): don't surface a stale summary while a newer run is active

### LINKED ISSUE #98204: [Bug]: Kanban drawer shows stale blocked summary as current while newer run is active
## Bug description

The Kanban task drawer can show a stale historical blocker as `LATEST SUMMARY` while the card header correctly says `RUNNING`. The comments section also keeps the old `BLOCKED:` message without an obvious historical/run label. Users reasonably interpret the task as still blocked even though a newer worker is active.

## Real reproduction

Task `t_68fa2185` had:

1. run 51 → `blocked`, summary: `Production blocked: ... ring vs track is only 1.36:1 ...`
2. corrected dependency completed
3. run 52 → running, later reclaimed
4. run 54 → currently running with live heartbeats

The task row/header was `RUNNING`, but the drawer displayed the run-51 blocker under `LATEST SUMMARY` and the historical `BLOCKED:` comment immediately below. Screenshot evidence showed both simultaneously.

## Root cause

At upstream `4209d371`, `hermes_cli/kanban_db.py::latest_summary()` intentionally selects the latest **non-empty** summary:

```sql
SELECT summary FROM task_runs
WHERE task_id = ? AND summary IS NOT NULL AND summary != ''
ORDER BY COALESCE(ended_at, started_at) DESC, id DESC LIMIT 1
```

A new active run has no summary yet, so the previous blocked run remains the top-level `latest_summary`. The dashboard renders it without source run ID, outcome, timestamp, or a historical label.

## Expected behavior

When the newest run is active and an older summary is surfaced for context, the UI should clearly label it, for example:

- `Previous run summary · Run 51 · Blocked · <timestamp>`; or
- suppress top-level `Latest Summary` during an active newer run and show historical summaries only in Run History.

Historical comments should remain immutable, but blocker comments should display their timestamp/run context so they cannot be mistaken for current task state.

## Actual behavior

`RUNNING` and an unlabeled stale `Production blocked...` summary appear together, making current lifecycle truth ambiguous.

## Suggested regression test

Create task → blocked run with summary → start newer running run with no summary → render drawer. Assert that the blocked text is either absent from the top-level summary or explicitly labeled as belonging to the prior blocked run, while current state remains `RUNNING`.

bsdnn:
## What does this PR do?

`latest_summary()` / `latest_summaries()` returned the newest run that *has* a
summary. A newer, still-active run has no summary yet, so an older blocked/
completed run's summary surfaced as the task's LATEST SUMMARY — a RUNNING task
then reads as still blocked, and the dashboard drawer shows a stale blocker as
current (#98204).

Fix: suppress a run's summary when a more recently created run is still active
(`ended_at IS NULL`), in both the single and batch summary queries. Once the
newer run finishes with its own summary, that summary surfaces as before.

Scope: top-level Latest Summary only. The drawer's DEPENDENCIES ("Blocked by"
vs "Depends on") labeling raised later in the same issue is left for a separate
follow-up.

Credit to @MK-Henry for the diagnosis in #98204.

## Related Issue

Fixes #98204

## Type of Change

## Changes Made

- `hermes_cli/kanban_db.py` — `latest_summary` and `latest_summaries` skip a
  stale summary while a newer run is still active.
- `tests/hermes_cli/test_kanban_db.py` — suppression + finished-run control tests.

## How to Test

    pytest tests/hermes_cli/test_kanban_db.py -k latest_summary -q

Both new tests pass: a stale blocked summary is suppressed while a newer run is
active; the newer run's own summary surfaces once it finishes.

## Checklist

- [√] I've read the Contributing Guide
- [√] My commit messages follow Conventional Commits
- [√] I searched for existing PRs / issues
- [√] My PR contains only changes related to this fix
- [√ ] I've run `pytest tests/ -q` and all tests pass
- [√] I've added tests for my changes
- [√] I've tested on my platform: Windows 11

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -12106,9 +12106,17 @@ def latest_summary(conn: sqlite3.Connection, task_id: str) -> Optional[str]:
     for ties or unfinished rows). Returns None if no run has a summary.
     """
     row = conn.execute(
-        "SELECT summary FROM task_runs "
-        "WHERE task_id = ? AND summary IS NOT NULL AND summary != '' "
-        "ORDER BY COALESCE(ended_at, started_at) DESC, id DESC LIMIT 1",
+        "SELECT summary FROM task_runs AS t "
+        "WHERE t.task_id = ? AND t.summary IS NOT NULL AND t.summary != '' "
+        # Suppress an older run's summary while a newer run is still active
+        # (unfinished, no summary of its own yet). The task's current state is
+        # that newer run, so surfacing the old summary as "latest" makes a
+        # RUNNING task read as still blocked/complete (#98204).
+        "AND NOT EXISTS ("
+        " SELECT 1 FROM task_runs AS newer "
+        " WHERE newer.task_id = t.task_id AND newer.id > t.id "
+        " AND newer.ended_at IS NULL) "
+        "ORDER BY COALESCE(t.ended_at, t.started_at) DESC, t.id DESC LIMIT 1",
         (task_id,),
     ).fetchone()
     return row["summary"] if row else None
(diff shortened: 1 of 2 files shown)


liyangbing:
<h3>Review</h3><p>This corrects the stale-summary contract in both single-task and batch queries: an older completed/blocked summary is no longer presented while a higher-id run for the same task is active. The finished-run control preserves the prior newest-summary behavior, and the <code>NOT EXISTS</code> predicate keeps the rule in SQL rather than relying on callers to infer run state.</p><h3>Suggested contract</h3><ul><li>Define “newer” explicitly. The suppression uses <code>newer.id &gt; t.id</code>, while selection orders by timestamps and id; document or test that the autoincrement id is the authoritative creation order and remains valid for imported/backfilled runs.</li><li>Extend the batch test with multiple task ids: one with an active newer run, one with only completed runs, one with an active run that already has a summary, and one with no summary. Verify there is no cross-task suppression.</li><li>Cover status transitions and empty values: an active run with an empty summary, a run whose <code>ended_at</code> is set but status remains running, and a newer run whose summary is written in the same transaction as completion.</li><li>Check the query plan/indexing on large task histories. The correlated <code>NOT EXISTS</code> predicate should have an index-supported path on <code>(task_id, id, ended_at)</code> or an equivalent existing index, especially for <code>latest_summaries</code>.</li><li>Verify snapshot/transaction semantics when a run starts between selecting summaries and rendering the dashboard; the UI should not mix an old summary with a newer run state from a different read snapshot.</li></ul><h3>Regression matrix</h3><ul><li>single and batch lookup with active newer run, completed newer run, no summary, and multiple task ids</li><li>older blocked, completed, failed, and cancelled summaries while the current run is active</li><li>active run with its own summary versus active run with empty/null summary</li><li>non-monotonic timestamps, imported rows, equal timestamps, and id tie-breaks</li><li>large history/query-plan coverage plus completion and dashboard reads in one transaction</li></ul><p>The focused suppression test and finished-run control are a good baseline. The batch/multi-task and creation-order cases would make the SQL contract robust against historical data and future run-import paths.</p>

PR #98211, opened 2026-08-30
fix(models): Anthropic /v1/models catalogs no longer stop at the first page (OpenHands port)

teknium1:
## Summary
Anthropic model catalogs no longer stop at the first page: Anthropic's `/v1/models` is cursor-paginated with a **default page size of 20**, and both hermes fetchers read one unpaginated page — every model past page 1 silently vanished from the `/model` picker and provider catalogs.

Root cause found scouting OpenHands (their frontend hit the identical bug against cloud provider catalogs and fixed it by following `next_page_id` — OpenHands/OpenHands#16758). Hermes' Anthropic readers have the same shape.

## Changes
- `hermes_cli/models.py` `_fetch_anthropic_models()`: request `limit=1000` and follow `has_more`/`last_id` with `after_id` (bounded to 20 pages, repeated-cursor guard, order-preserving de-dupe)
- `plugins/model-providers/anthropic/__init__.py` `fetch_models()`: same pagination walk; also now honors its `base_url` argument instead of hardcoding `api.anthropic.com`
- `tests/hermes_cli/test_anthropic_models_pagination.py` (new): real-HTTP paginated mock server (page cap 20, Anthropic cursor semantics) covering both fetchers — full walk, single-page catalog, stuck-cursor termination
- `tests/hermes_cli/test_anthropic_pool_model_discovery.py`: URL pins updated for the `?limit=1000` contract

## Validation
| Fetcher (server serves 55 models, 20/page) | Before | After |
|---|---|---|
| `_fetch_anthropic_models` | 20 models | 55 models |
| plugin `fetch_models` | `None` (base_url ignored → hardcoded host) | 55 models |

**Live repro:** paginated mock server on `origin/main` → both fetchers fail as above; with the fix both return 55/55. Targeted tests: 10/10 pass (`test_anthropic_models_pagination.py`, `test_anthropic_pool_model_discovery.py`, `test_anthropic_picker_curated.py`).

Non-Anthropic siblings checked: OpenAI-compatible `/v1/models` catalogs (OpenRouter, Nous, etc.) return full lists unpaginated; the generic custom-catalog cursor case is separately covered by open PR #96781 (`next_cursor` shape) — different wire contract, no overlap with Anthropic's `after_id` walk.

Source: weekly OpenHands PR scout (cron).

## Infographic

Anthropic catalog pagination fix


liyangbing:
<h3>Review</h3><p>This closes the real catalog truncation in both Anthropic fetchers: it requests a larger page, follows <code>has_more</code>/<code>last_id</code> with <code>after_id</code>, guards repeated cursors, and de-duplicates without losing order. Fixing the plugin's ignored <code>base_url</code> at the same time is important for proxy/pool deployments, and the mock server tests exercise the two fetchers independently.</p><h3>Suggested contract</h3><ul><li>Verify that every paginated request preserves the provider headers and the caller's resolved endpoint, including a base URL with a path or existing query parameters; add assertions for the requested cursor sequence and auth headers on page 2+.</li><li>Define malformed-response behavior separately from a normal end: non-boolean <code>has_more</code>, missing/empty <code>last_id</code>, duplicate model ids, malformed <code>data</code>, and a cursor that advances but returns the same page should terminate safely and expose a diagnostic rather than silently presenting a partial catalog.</li><li>Confirm the bounded page limit is intentional for catalogs larger than 20,000 models and that hitting the 20-page cap is observable. A silent partial result can make the picker look complete while omitting models.</li><li>Test retry and transient failure on a later page. The first page succeeding must not cause a later page error to be mistaken for a complete catalog, and the two fetchers should have consistent failure semantics.</li><li>Preserve deterministic picker ordering after cross-page de-duplication, including duplicate ids with different metadata and ids that do not match the usual Claude naming pattern.</li></ul><h3>Regression matrix</h3><ul><li>one page, multiple pages, exactly-at-cap, and beyond-cap catalogs</li><li>cursor sequence, repeated cursor, empty cursor, duplicate ids, reordered pages, and malformed page payloads</li><li>default Anthropic endpoint, custom base URL, proxy path, existing query string, and page-2 header verification</li><li>first-page and later-page HTTP/JSON failures, timeout, retry, and partial-result policy</li><li>picker sort/dedup behavior with mixed model ids and a catalog larger than the default page size</li></ul><p>The reported 55/55 live mock result and 10 focused passes are strong evidence. Request-sequence/header assertions and an explicit beyond-20-page policy would make the pagination contract safer for real provider catalogs and proxy deployments.</p>

PR #9795, opened 2026-04-14
feat(wecom): thinking indicator + native stream consumer for reasoning & visible text streaming

agentlinker:
## What

Add thinking indicator and reasoning & visible text streaming support for the WeCom platform.

## Why

WeCom does not support editing sent messages, so the existing edit-based `GatewayStreamConsumer` does not work well for progressive output. WeCom provides a native stream message API (`replyStream`) that supports incremental updates via think tags, which makes it possible to implement a first-class thinking indicator and token streaming experience.

### New file: `gateway/wecom_stream_consumer.py`
- Add a WeCom-native stream consumer built on `replyStream` instead of edit-based message updates
- Receive visible tokens via `on_delta()`
- Reasoning tokens via `on_reasoning()`
- Send progressive updates through WeCom stream messages
- Rotate the active stream every 5 minutes to stay below WeCom's stream lifetime limit (~6 min)
- Cancel the pre-stream thinking loop only after the first visible/reasoning token arrives
- Preserve the reasoning block in the final message when the model returns reasoning (format: `<think>...</think>`)
- **[Updated]** Add `on_error()` callback for graceful stream closure with error hint when API calls fail mid-stream
- **[Updated]** Add `STREAM_MAX_CONTENT_LENGTH` (3500 chars) rotation to prevent SDK from silently splitting long messages across frames
- **[Updated]** Add `error_text` parameter to `build_ws_stream_content()`

### Modified: `gateway/platforms/wecom.py`
- Mark WeCom as `SUPPORTS_MESSAGE_EDITING = False`
- Implement a 1-second thinking loop that emits `等待模型响应 Ns` lines using think tags, matching OpenClaw's `buildWaitingModelContent` behavior
- Keep the thinking loop alive until the first visible/reasoning token arrives
- Add configurable safety timeout via `HERMES_THINKING_MAX_SECONDS` (default: 1800s / 30 min)

### Modified: `gateway/display_config.py`
- Scope streaming resolution change to WeCom only
- When there is no explicit `display.platforms.wecom.streaming` override, `resolve_display_setting(..., "streaming")` now returns `None` (follow top-level `streaming.*` config)
- Other low-tier platforms (Signal, DingTalk, etc.) retain their existing default `False` behavior from tier defaults
- This makes top-level `streaming.*` the source of truth **for WeCom streaming**, instead of low-tier platform defaults like `_TIER_LOW["streaming"] = False`

### Modified: `gateway/run.py`
[...]

PR #98212, opened 2026-08-30
fix(gateway): keep provider stall-watchdog kill notices out of chat

agentlinker:
## Summary

`_buffer_status` records the provider stall-watchdog kill notices in the retry buffer. On terminal failure `_flush_status_buffer` replays them through `_emit_status` → `status_callback` → `_prepare_gateway_status_message`, so they land in a chat bubble on every messaging surface:

```
⚠️ No response from provider for 150s (non-streaming, model: gpt-5.5). Aborting call.
⚠️ No first byte from provider in 120s (codex stream, model: gpt-5.5). Reconnecting.
⚠️ Codex stream sent no events for 90s after first byte (model: gpt-5.5). Reconnecting.
⚠️ No events from Bedrock for 300s (model: anthropic...). Aborting...
```

The retry loop reconnects on its own and the user already gets a `⏳ Working — N min` heartbeat, so the raw timeout/context-size diagnostics are operator noise on a chat surface. Every other notice from the same retry ladder is already suppressed by `_TELEGRAM_NOISY_STATUS_RE` — `Rate limited. Waiting Ns`, `Retrying in Ns`, `Max retries (N) exhausted`, `stream drop … retry N`. These four were the ones left through.

**Whole family, not a subset.** Which shape fires depends only on *when* the provider goes silent — before the first byte, mid-stream, non-streaming, or Bedrock. Covering a subset would make chat noise depend on provider timing: a Codex user hitting the TTFB cutoff would get silence, while the same user hitting the post-TTFB idle cutoff a second later would get a bubble.

## Changes

- `gateway/run.py` (+11/-2): four alternatives in `_TELEGRAM_NOISY_STATUS_RE`, one per emit site family. The mandatory `\d+s` duration keeps ordinary prose about providers and timeouts from matching.
- `tests/gateway/test_telegram_noise_filter.py`: shared `PROVIDER_WATCHDOG_KILL_NOTICES` / `BUFFERED_STATUS_MUST_SURVIVE` constants, suppression asserted across 4 chat platforms, plus a positive guard on the notices that share the same status path.
- `tests/gateway/test_provider_watchdog_status_delivery.py` (new, +130): end-to-end buffer→surface delivery test.

Emit sites covered:

| Notice | Emit site |
|---|---|
| `No response from provider for Ns` (non-streaming stale kill) | `agent/chat_completion_helpers.py:787` |
| `No response from provider for Ns` (streaming stale kill) | `agent/chat_completion_helpers.py:5211` |
| `No first byte from provider in Ns` (codex TTFB kill) | `agent/chat_completion_helpers.py:1670`, `:1676` |
| `Codex stream sent no events for Ns after first byte` (codex post-TTFB idle kill) | `agent/chat_completion_helpers.py:1725` |
| `No events from Bedrock for Ns` (bedrock stream stale kill) | `agent/chat_completion_helpers.py:3615` |

Programmatic surfaces (local/api/webhook/msgraph_webhook) are unaffected — they return before the filter via `_gateway_surface_passes_raw_text`. `test_programmatic_surfaces_keep_raw_status` is parametrized over every watchdog notice, so narrowing the chat filter can never quietly strip a timeout diagnostic from an API/webhook consumer.

## Notices that must survive the same path

The filter sits on the path that also carries notices the user needs. Each now has an explicit guard on all 4 chat platforms — a widened alternative that swallowed one of these would turn a failed turn into a turn that just stops replying:

| Notice | Emit site | Why it must reach chat |
|---|---|---|
| `❌ Connection to provider failed after N attempts…` | `chat_completion_helpers.py:5026` | terminal, user-actionable |
| `⚠️ Model fallback: X via Y unavailable…` | `chat_completion_helpers.py:2856` | durable state change |
| `⏳ Working — N min` | `gateway/run.py:29992` | work heartbeat |

## Test design note

An earlier revision of this PR had a negative test feeding ordinary assistant prose to `_prepare_gateway_status_message`. That test asserted nothing: the function's only caller is `_status_callback_sync` (`gateway/run.py:5426`), and `status_callback` is only ever invoked with `lifecycle` / `warn` / `compacted` — never assistant prose, which travels `_sanitize_gateway_final_response` instead. It has been replaced with the reachable negative cases in the table above.

`tests/gateway/test_provider_watchdog_status_delivery.py` drives the real chain with only the adapter sink stubbed:

```
_buffer_status → _flush_status_buffer → _emit_status
  → status_callback("lifecycle", …) → _prepare_gateway_status_message (real) → sink
```

It asserts four things: the notice reaches stdout/logs but produces no chat bubble (suppression, not deletion); a mixed terminal ladder delivers exactly the actionable subset in order; successful recovery (`_clear_status_buffer`) emits nothing on either surface; and `api_server` still receives the raw text through the same chain.

## Scope note

The same retry ladder also emits `API call failed (attempt N/M)`, `🔌 Provider:`, `🌐 Endpoint:`, `📝 Error:`, `⏱️ Elapsed:` and `💀 Final error:`. Those go through `_buffer_vprint` / `_vprint`, which `_flush_status_buffer` routes to `_vprint` → `_safe_print` → `_print_fn` (default `print`) → process stdout. The gateway never reassigns `_print_fn` for a session agent and never redirects stdout, so they cannot reach `_prepare_gateway_status_message`. Filtering them would be dead code, so this PR deliberately leaves them alone.

## Validation

| Check | Result |
|---|---|
| noise filter + delivery + `tests/run_agent/test_retry_status_buffer.py` | 270 passed, 0 failed |
| same files with only the 2 sibling regex alternatives reverted | 20 failed, 239 passed |
| `test_compression_progress_notices` + `test_turn_context_overflow_warning` + `test_compression_count_warning_36908` | 49 passed |
| `ruff check` on all 3 changed files | clean |
| `tests/gateway/` full run | 6727 passed, 26 failed, 18 skipped |

The 26 gateway failures are pre-existing on `main` (`delivery_ledger`, `whatsapp_bridge_pidfile`, `wecom_callback`, `scale_to_zero`, …). None of the 26 failing files reference `_prepare_gateway_status_message`, `_TELEGRAM_NOISY_STATUS_RE`, or any watchdog notice text (`grep -c` = 0 for each), so none are caused by this change.

## Context

Extracted from #9795 (closed as superseded by #96425). That branch bundled WeCom native streaming with unrelated fixes; this is the first of the focused resubmissions requested in the closing note. The WeCom-transport parts of #9795 are dropped — #96425 supersedes them.

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -148,13 +148,19 @@ _TELEGRAM_NOISY_STATUS_RE = re.compile(
     r"|max\s+retries\s+\(\d+\).*(?:trying\s+fallback|exhausted|invalid\s+responses)"
     r"|stream\s+(?:drop|drop\s+mid\s+tool-call).+retry\s+\d"
     r"|stale\s+connections\s+from\s+a\s+previous\s+provider\s+issue"
-    # Provider stale/TTFB kill notices buffered by _buffer_status in
+    # Provider stall-watchdog kill notices buffered by _buffer_status in
     # agent/chat_completion_helpers.py and replayed through _emit_status when
     # a turn exhausts its attempts. The retry loop reconnects on its own, so
     # these belong in logs, not in a chat bubble. The mandatory "\d+s" duration
     # keeps ordinary assistant prose about providers/timeouts from matching.
-    r"|no\s+response\s+from\s+provider\s+for\s+\d+s\b"
-    r"|no\s+first\s+byte\s+from\s+provider\s+in\s+\d+s\b"
+    # All four shapes of the same watchdog family are listed together: which
+    # one fires depends only on WHEN the provider goes silent (before the first
+    # byte, mid-stream, non-streaming, Bedrock), so covering a subset would
+    # make chat noise depend on provider timing.
+    r"|no\s+response\s+from\s+provider\s+for\s+\d+s\b"  # (non-)streaming stale kill
+    r"|no\s+first\s+byte\s+from\s+provider\s+in\s+\d+s\b"  # codex TTFB kill
+    r"|codex\s+stream\s+sent\s+no\s+events\s+for\s+\d+s\b"  # codex post-TTFB idle kill
+    r"|no\s+events\s+from\s+bedrock\s+for\s+\d+s\b"  # bedrock stream stale kill
     rf"|{re.escape(COMPACTION_DONE_STATUS)}"
     r")",
     re.IGNORECASE | re.DOTALL,
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -148,6 +148,13 @@ _TELEGRAM_NOISY_STATUS_RE = re.compile(
     r"|max\s+retries\s+\(\d+\).*(?:trying\s+fallback|exhausted|invalid\s+responses)"
     r"|stream\s+(?:drop|drop\s+mid\s+tool-call).+retry\s+\d"
     r"|stale\s+connections\s+from\s+a\s+previous\s+provider\s+issue"
+    # Provider stale/TTFB kill notices buffered by _buffer_status in
+    # agent/chat_completion_helpers.py and replayed through _emit_status when
+    # a turn exhausts its attempts. The retry loop reconnects on its own, so
+    # these belong in logs, not in a chat bubble. The mandatory "\d+s" duration
+    # keeps ordinary assistant prose about providers/timeouts from matching.
+    r"|no\s+response\s+from\s+provider\s+for\s+\d+s\b"
+    r"|no\s+first\s+byte\s+from\s+provider\s+in\s+\d+s\b"
     rf"|{re.escape(COMPACTION_DONE_STATUS)}"
     r")",
     re.IGNORECASE | re.DOTALL,
--- a/tests/gateway/test_provider_watchdog_status_delivery.py
+++ b/tests/gateway/test_provider_watchdog_status_delivery.py
@@ -1,0 +1,136 @@
+"""End-to-end buffer→surface delivery for provider stall-watchdog notices.
+
+``tests/gateway/test_telegram_noise_filter.py`` asserts the filter *predicate*.
+This module drives the REAL path a notice travels on terminal failure:
+
+    agent._buffer_status(...)          # agent/chat_completion_helpers.py
+      -> agent._flush_status_buffer()  # run_agent.py, on retry exhaustion
+        -> agent._emit_status(...)     # run_agent.py
+          -> agent.status_callback("lifecycle", ...)
+            -> _prepare_gateway_status_message(...)   # gateway/run.py (real)
+              -> adapter.send(...)                    # stubbed sink
+
+Only the adapter sink is stubbed, so a regression that keeps the regex green
+while changing what actually reaches a chat bubble fails here.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from gateway.run import _prepare_gateway_status_message
+from run_agent import AIAgent
+from tests.gateway.test_telegram_noise_filter import (
+    BUFFERED_STATUS_MUST_SURVIVE,
+    CHAT_PLATFORMS,
+    PROVIDER_WATCHDOG_KILL_NOTICES,
+)
+
+
+def _agent_wired_to_surface(platform: str):
+    """A bare AIAgent whose status_callback is the real gateway filter.
+
+    Mirrors ``gateway/run.py:_status_callback_sync``: prepare the message, drop
+    it when the filter returns None, otherwise hand it to the adapter. Returns
+    ``(agent, bubbles, stdout_lines)`` — ``bubbles`` is what a chat user would
+    see, ``stdout_lines`` is the operator-facing CLI/log stream.
+    """
+    agent = object.__new__(AIAgent)
+    agent.log_prefix = ""
+    agent.suppress_status_output = False
+    agent._mute_post_response = False
+    agent._executing_tools = False
+
+    bubbles: list[str] = []
+    stdout_lines: list[str] = []
+    agent._print_fn = lambda *args, **kwargs: stdout_lines.append(
+        " ".join(str(a) for a in args)
+    )
+
+    def _status_callback(event_type: str, message: str) -> None:
+        prepared = _prepare_gateway_status_message(platform, event_type, message)
+        if prepared is None:
+            return
+        bubbles.append(prepared)
+
+    agent.status_callback = _status_callback
+    return agent, bubbles, stdout_lines
+
+
+@pytest.mark.parametrize("platform", CHAT_PLATFORMS)
+@pytest.mark.parametrize(
+    "notice", PROVIDER_WATCHDOG_KILL_NOTICES, ids=lambda m: m.strip()[:48]
+)
+def test_terminal_failure_keeps_watchdog_notice_out_of_chat(platform, notice):
+    """A flushed watchdog kill notice reaches stdout/logs but never a bubble."""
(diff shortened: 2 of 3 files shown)


liyangbing:
<h3>Review</h3><p>This is a focused transport-surface fix: the two provider stale/TTFB notices are buffered lifecycle diagnostics, and the patch suppresses them only when preparing chat-surface status messages while keeping programmatic surfaces raw. Adding WeCom to the chat-platform matrix is consistent with that boundary.</p><h3>Suggested contract</h3><ul><li>Keep the suppression decision tied to the lifecycle/status path, not merely the text. Add an integration-level assertion that a terminal retry emits the notice to logs/diagnostics but produces no chat bubble on each supported chat surface.</li><li>Strengthen the negative cases around the new regex: ordinary assistant prose containing the exact phrase plus a duration, different punctuation/casing, multiline text, and a model or provider name that happens to contain similar words.</li><li>Document whether the filter is intentionally applied to all four chat platforms while local/API/webhook/Microsoft Graph surfaces retain the raw text; test both sides through <code>_gateway_surface_passes_raw_text</code>.</li><li>Verify that both streaming and non-streaming notices are suppressed after retries are exhausted, while the existing heartbeat and final actionable error remain visible and are not accidentally swallowed.</li></ul><h3>Regression matrix</h3><ul><li>non-streaming stale kill, streaming stale kill, and Codex TTFB kill across Telegram, Slack, Feishu, and WeCom</li><li>retry reconnect followed by terminal failure, including the buffered notice flush path</li><li>local, API, webhook, and Microsoft Graph programmatic surfaces preserving raw diagnostics</li><li>exact-duration prose, multiline/case variants, and adjacent heartbeat/final-error messages</li><li>the pre-existing gateway test subset, with the documented baseline failures separated from this change</li></ul><p>The reported focused green result and red-before/green-after regression are useful evidence; exercising the actual buffer-to-surface path would close the remaining risk that the regex tests pass while delivery behavior changes elsewhere.</p>

agentlinker:
Thanks — point 4 was the useful one, and chasing it found a real gap. Pushed as e024b1a. Going through each suggestion with what I verified:

### The gap your review found

Not directly, but asking for the buffer-to-surface path to be exercised is what surfaced it: **two more notices from the same watchdog family were still reaching chat**, and the original PR body's claim that only two shapes were left through was wrong.

| Notice | Emit site | Was |
|---|---|---|
| `⚠️ Codex stream sent no events for {N}s after first byte …` | `agent/chat_completion_helpers.py:1725` | leaking |
| `⚠️ No events from Bedrock for {N}s … Aborting...` | `agent/chat_completion_helpers.py:3615` | leaking |

Both travel the identical `_buffer_status` → `_flush_status_buffer` → `_emit_status` → `status_callback` path as the two already covered, and are the same pure-diagnostic `\d+s` shape. Which one fires depends only on *when* the provider goes silent, so covering a subset made chat noise depend on provider timing — a Codex user hitting the TTFB cutoff got silence, the same user hitting the post-TTFB idle cutoff a second later got a bubble. Both are now in the regex, and the PR body/title are corrected to say 5 emit sites rather than 2.

### 1. "Keep the suppression decision tied to the lifecycle/status path, not merely the text"

It already is, structurally. `_prepare_gateway_status_message` has exactly one caller — `_status_callback_sync` (`gateway/run.py:5426`) — and `status_callback` is only ever invoked with three event types repo-wide:

```
$ grep -rhno 'status_callback(\s*"[a-z_]*"' --include='*.py' . | grep -o '"[a-z_]*"' | sort | uniq -c
   4 "lifecycle"
   2 "warn"
   1 "compacted"
```

Assistant prose never reaches this function; it travels `_sanitize_gateway_final_response`. So the text match is already scoped to the status path, and gating additionally on `event_type` would be a no-op — all three types are equally operator-facing.

### 2. "An integration-level assertion that a terminal retry emits to logs/diagnostics but produces no chat bubble"

Added, and this is the substantive change on the test side. New `tests/gateway/test_provider_watchdog_status_delivery.py` drives the real chain with only the adapter sink stubbed:

```
_buffer_status → _flush_status_buffer → _emit_status
  → status_callback("lifecycle", …) → _prepare_gateway_status_message (real) → sink
```

Four assertions, each × 6 notice shapes × 4 chat platforms:
- notice lands on stdout/logs, **zero** chat bubbles — suppression, not deletion
- nothing emits while the buffer is still filling (only on flush)
- a mixed terminal ladder (2 watchdog kills → fallback switch → stale kill → terminal give-up) delivers exactly `[fallback, terminal]`, in order, while all 5 stay on the operator surface
- successful recovery (`_clear_status_buffer`) emits nothing on either surface
- `api_server` receives the raw text through the same chain

### 3. "Strengthen the negative cases: prose with the exact phrase plus a duration, punctuation/casing, multiline"

I'd push back on most of this, but it pointed at a real problem with the test I had.

Casing and multiline are already covered by construction — the pattern is `re.IGNORECASE | re.DOTALL`, and I confirmed `no response from provider for 45S` and a `\n`-separated variant both match. Prose containing the exact phrase *plus* a duration is a true positive by design, not a case that should survive.

But checking this exposed that my negative test was **vacuous**: it fed assistant prose to a function that (per point 1) can never receive assistant prose. Removed. The reachable negative cases are the notices that share the status path, which now have explicit guards on all 4 chat platforms:

| Must survive | Emit site | Why |
|---|---|---|
| `❌ Connection to provider failed after N attempts…` | `chat_completion_helpers.py:5026` | terminal, user-actionable |
| `⚠️ Model fallback: X via Y unavailable…` | `chat_completion_helpers.py:2856` | durable state change |
| `⏳ Working — N min` | `gateway/run.py:29992` | work heartbeat |

A widened alternative that swallowed any of these would turn a failed turn into one that just stops replying — that's the regression worth locking, and it wasn't locked before.

### 4. "Verify both notices are suppressed after exhaustion, while heartbeat and final actionable error remain visible"

This was correct and previously untested. Verified across every buffered notice shape:

```
SUPPRESSED | cch:787  non-streaming stale kill
SUPPRESSED | cch:5211 streaming stale kill
SUPPRESSED | cch:1670 codex TTFB kill
SUPPRESSED | cch:1725 codex post-TTFB idle kill   ← newly covered
SUPPRESSED | cch:3615 bedrock stream stale kill   ← newly covered
SUPPRESSED | stream_diag:261 stream drop retry
--- must survive ---
DELIVERED  | cch:5026 terminal give-up
DELIVERED  | cch:2856 model fallback switch
DELIVERED  | run:29992 heartbeat
DELIVERED  | cl:6207  safety filter → fallback
DELIVERED  | cl:5519  billing exhausted → fallback
```

### 5. Documenting the chat-vs-programmatic split

PR body now has the emit-site table and the must-survive table. `test_programmatic_surfaces_keep_raw_status` is parametrized over all 6 watchdog notices × 4 programmatic surfaces (it previously used one unrelated retry string), so narrowing the chat filter can't quietly strip a timeout diagnostic from an API/webhook consumer.

Both notice lists are hoisted to module level and shared by the predicate test, the programmatic-surface test, and the delivery test — a new emit site gets covered everywhere by editing one list.

### Validation

| Check | Result |
|---|---|
| noise filter + delivery + `test_retry_status_buffer.py` | 270 passed, 0 failed |
| same files, only the 2 new regex alternatives reverted | **20 failed**, 239 passed |
| `ruff check` on all 3 changed files | clean |
| `tests/gateway/` full run | 6727 passed, 26 failed, 18 skipped |

The 26 are pre-existing on `main`. `grep -c` for `_prepare_gateway_status_message`, `_TELEGRAM_NOISY_STATUS_RE`, and every watchdog notice string returns 0 in all 26 failing files, so none are caused by this change.

Issue #15759, opened 2026-04-25
[Bug]: read_file dedup stub never triggers loop detection — infinite read loop


strueman:
### Bug Description

When using Qwen3.6 models hosted locally, the agent occasionally gets stuck in an infinite loop when troubleshooting code. This is a new issue with the same models that worked previously. I have tested multiple versions and quants from Unsloth, Bartowski of Qwen3.6 35b A3B and 27b. The only thing that changed in the setup before this started happening was running hermes update, after troubleshooting the updated instance I installed a fresh version of the latest main branch in a vm and had the same results with long running tasks involving file reading.

### Steps to Reproduce

1. Model calls `read_file` on a file → gets content
2. Model calls `read_file` on the same file again → gets dedup stub:
   ```json
   {"content": "File unchanged since last read. The content from the earlier read_file result in this conversation is still current — refer to that instead of re-reading.", "path": "...", "dedup": true}
   ```
3. Model calls `read_file` again → **same dedup stub** (infinite loop)

The existing loop-detection guard at `tools/file_tools.py:538` (hard block after 4 consecutive reads) and the warning at line 549 (after 3 consecutive reads) are **never reached** for dedup stubs because the dedup path returns early at line 447.

### Expected Behavior

```
read_file → content      (dedup mtime cached)
read_file → dedup stub   (dedup_hits[key]=1)
read_file → BLOCKED      (dedup_hits[key]=2 → hard error: "STOP calling read_file")
```

### Actual Behavior

```
read_file → content      (consecutive=1)
read_file → dedup stub   (dedup path returns early, never hits loop detection)
read_file → dedup stub   (same stub forever)
...
```

### Affected Component

Agent Core (conversation loop, context compression, memory)

### Debug Report

```shell
2026-04-26 04:44:26,890 INFO hermes_cli.plugins: Plugin 'openai' registered image_gen provider: openai
2026-04-26 04:44:26,891 INFO hermes_cli.plugins: Plugin 'openai-codex' registered image_gen provider: openai-codex
2026-04-26 04:44:26,921 INFO hermes_cli.plugins: Plugin 'xai' registered image_gen provider: xai
2026-04-26 04:44:26,959 INFO hermes_cli.plugins: Plugin 'hermes-lcm' registered context engine: lcm
2026-04-26 04:44:26,959 INFO hermes_plugins.hermes_lcm: LCM slash command registration disabled (set LCM_ENABLE_SLASH_COMMAND=1 to enable /lcm)
[...]

PR #98213, opened 2026-08-30
fix(gateway): scope out-of-turn compression dedup reset to the live session task

### LINKED ISSUE #98206: [Bug]: skill_view can return a dedup stub after compression prunes the original skill content
## Bug Description

After context compression replaces a previously loaded skill with `[SKILL_PRUNED]`, a subsequent `skill_view` call can return the unchanged-content deduplication stub instead of the skill text:

```json
{
  "success": true,
  "status": "unchanged",
  "dedup": true,
  "content_returned": false,
  "message": "Skill content unchanged since it was loaded earlier in this conversation — refer to the earlier skill_view result; it is still current and complete. (Re-issued after context compression, this returns the full content again.)"
}
```

The referenced earlier result is no longer available to the model. This can make required skill instructions inaccessible while telling the model they remain loaded, causing instruction-compliance failures.

This appears to violate the post-compression invariant documented in PR #77095. It is the `skill_view` analogue of the task-scoped reset problem being investigated in #90949.

## Steps to Reproduce

Observed in a live Telegram gateway conversation:

1. Load a large skill with `skill_view` and continue the conversation until context compression replaces that tool result with `[SKILL_PRUNED: content lost in compression; reload with skill_view(name='…')]`.
2. Call `skill_view` for that same skill.
3. Observe another `[SKILL_PRUNED]` result or otherwise continue and call `skill_view` again.
4. Observe a deduplication response with `content_returned: false` rather than the complete `SKILL.md` content.

The live reproduction used Hermes Agent v0.20.6 at commit `692c2f9d6c3a72b498a653ac0cc052f311c4c728`. At filing time, upstream `main` is `4209d371aa1bb8840ce8447555bdd863a1a96c38`; the current source still relies on task-scoped `reset_skill_view_dedup(task_id)` during compression.

A durable regression should exercise the real gateway compaction entry point with a non-default live task ID rather than calling `reset_skill_view_dedup()` directly:

1. Load a skill under a non-default task ID.
2. Trigger the gateway's actual context-compression path.
3. Confirm the original skill result is no longer present in live model context.
4. Request the skill again.
5. Assert that complete content is returned.
6. Assert that another task's dedup state remains isolated.

## Expected Behavior

A `skill_view` dedup stub is valid only while the complete referenced skill content remains available to the model. After context compression evicts or summarizes that content, the next `skill_view` request must return the complete content.

## Actual Behavior

The post-compression request can return `dedup: true` and `content_returned: false`, referring to content that has been removed from the model-visible context.

## Affected Component

- Gateway (Telegram/Discord/Slack/WhatsApp)

## Messaging Platform

Telegram

## Debug Report

A public debug-report upload was intentionally omitted because the full report includes unrelated private gateway/session material. The sanitized environment, exact tool response, source commit, and reproduction are included above.

## Operating System

macOS 26.5.2 (25F84)

## Python Version

3.11.15 in the Hermes environment

## Hermes Version

Hermes Agent v0.20.6 (2026.8.27), commit `692c2f9d6c3a72b498a653ac0cc052f311c4c728`

## Root Cause Analysis

**Hypothesis, not yet confirmed:** the gateway hygiene-compaction path may reset the dedup tracker under a different or default task ID instead of the live conversation's task ID. Issue #90949 identifies this possible lifecycle mismatch for `read_file`; `skill_view` is reset alongside it in `agent/conversation_compression.py`.

PR #94907 adds a task-scoped reset regression, but its test directly calls `reset_skill_view_dedup("compressed-task")`. That proves the helper works when supplied the correct ID; it does not exercise the gateway compaction path or prove that the live task ID reaches the helper.

## Proposed Fix

Trace the task ID used when recording the initial `skill_view`, the task ID supplied during gateway compression, and the task ID used for the post-compression read. If they differ, propagate the live task ID through the gateway compaction path and cover both `skill_view` and `read_file` with end-to-end regressions.

Do not persist a content-free dedup reference across compression. The invariant should be availability-based: if the referenced content is no longer in model-visible context, return the complete content.

Related: #77095, #90949, #84857, #44819, and #94907.

liuhao1024:
## What does this PR do?

Fixes a post-compression lifecycle defect where `skill_view` (and `read_file`) could return an unchanged-content dedup stub referencing content that context compression had already pruned from model-visible context.

Root cause: the gateway's two out-of-turn compression entry points — the manual `/compress` command (`gateway/slash_commands.py`) and the session-hygiene sweep (`gateway/run.py`) — call `_compress_context` **without** `task_id`, so the dedup resets inside `compress_context` (`reset_skill_view_dedup` / `reset_file_dedup`) fall back to the "default" task scope. But live tool reads are keyed by `session_entry.session_id` — the same value the main gateway turn passes to `run_conversation` as its `task_id` (gateway/run.py: `_run_agent(session_id=session_entry.session_id)` → `TurnContext` → `task_id=ctx.session_id`). The fallback reset therefore clears nothing for the live session, and the next `skill_view` for an unchanged skill returns `content_returned: false` pointing at a `[SKILL_PRUNED]` placeholder.

This PR forwards `session_entry.session_id` on both entry points, scoping the reset to the same task key the live turn records dedup state under.

Note on the task key: it must be the session **row id** (`session_entry.session_id`, e.g. `20260829_231501_ab12cd34` per `_get_or_create_session_impl`), not the session *key* (`agent:main:telegram:group:...`). The session key never reaches `run_conversation`'s `task_id` on the live path, so resetting under it would miss the live records in the same way "default" does.

Both compression outcomes stay correct: on an **in-place** compaction the session id is unchanged, so the reset now actually clears the live records (this is the fix); on a **rotated** compaction the reset clears the old id's records (harmless) while the next turn starts under a fresh id that has no records to hit.

Credit: the task-ID hypothesis in #98206 was independently confirmed by @0xble (who could not open a cross-repo PR); this PR forwards `session_entry.session_id` rather than the session key for the reason above.

## Related Issue

Fixes #98206

## Type of Change

## Changes Made

- `gateway/run.py` — hygiene sweep: pass `task_id=session_entry.session_id` to `_hyg_agent._compress_context` (was falling back to "default").
- `gateway/slash_commands.py` — `/compress` handler: pass `task_id=session_entry.session_id` to `tmp_agent._compress_context` (same fallback).
- `tests/gateway/test_compress_command.py` — extend the existing full-path `/compress` regression to assert the live session id rides along as `task_id`.
- `tests/gateway/test_session_hygiene.py` — extend the existing full-path hygiene regression (`_handle_message`) to capture and assert the forwarded `task_id`.

## How to Test

1. `python3 -m pytest tests/gateway/test_compress_command.py tests/gateway/test_session_hygiene.py tests/tools/test_skill_view_dedup.py tests/gateway/test_compress_focus.py -q` — Observed result: **34 passed**.
2. Fail-before check: with the two source changes stashed (tests kept), both regressions fail on `assert kwargs.get('task_id') == 'sess-1'` / the captured-task assertion; with the fix restored they pass.
3. `ruff check gateway/run.py gateway/slash_commands.py tests/gateway/test_compress_command.py tests/gateway/test_session_hygiene.py` — Observed result: All checks passed.

### Code

- [x] I've tested on my platform: macOS 26.5 (arm64)

## Screenshots / Logs

Not applicable (behavioral fix covered by the regressions above).

### DIFF
--- a/agent/conversation_compression.py
+++ b/agent/conversation_compression.py
@@ -2721,7 +2721,16 @@ def compress_context(
         system_message: Current system prompt; used when compression needs a
             rebuilt cached prompt.
         approx_tokens: Pre-compression token estimate, logged for ops.
-        task_id: Tool task scope (used for clearing file-read dedup state).
+        task_id: Tool task scope the post-compaction dedup reset applies to
+            (``reset_file_dedup``/``reset_skill_view_dedup``). Gateway callers
+            MUST pass the live session row id — ``session_entry.session_id``,
+            the same task_id the main turn hands to ``run_conversation``: the
+            live turn records its tool-read dedup state under that id, and
+            resetting any other scope leaves those records pointing at pruned
+            context (#98206). ``None`` is not a valid scope here: both resets
+            treat it as "clear every task", which breaks concurrent-session
+            isolation. The ``"default"`` fallback is only correct for
+            non-gateway callers that have no per-session task scope.
         focus_topic: Optional focus string for guided compression — the
             summariser will prioritise preserving information related to
             this topic.  Inspired by Claude Code's ``/compact <focus>``.
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -20487,6 +20487,18 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                                             _hyg_msgs, "",
                                             approx_tokens=_approx_tokens,
                                             commit_fence=_hyg_commit_fence,
+                                            # The live turn's tool calls (and
+                                            # thus the skill_view/read_file dedup
+                                            # records) are scoped to the session
+                                            # row id — the same value the main
+                                            # turn passes as run_conversation's
+                                            # task_id. Reset under that key so a
+                                            # post-compression re-read returns
+                                            # full content instead of a dedup
+                                            # stub pointing at pruned context
+                                            # (#98206); the "default" task
+                                            # fallback clears nothing here.
+                                            task_id=session_entry.session_id,
                                         ),
                                     )
                                     try:
--- a/gateway/slash_commands.py
+++ b/gateway/slash_commands.py
@@ -4658,6 +4658,13 @@ class GatewaySlashCommandsMixin:
                         focus_topic=focus_topic,
                         force=True,
                         defer_context_engine_notification=True,
+                        # Scope the dedup reset to the live session row id —
+                        # the same task_id the main gateway turn passes to
+                        # run_conversation. Without it the reset falls back to
+                        # the "default" task and a post-compression skill_view
+                        # can return a dedup stub pointing at pruned context
+                        # (#98206).
+                        task_id=session_entry.session_id,
                     )
                 )
 
--- a/tests/gateway/test_compress_command.py
+++ b/tests/gateway/test_compress_command.py
@@ -102,6 +102,14 @@ async def test_compress_command_works_when_auto_compaction_disabled():
     assert "Compressed:" in result
     agent_instance._compress_context.assert_called_once()
     assert agent_instance._compress_context.call_args.kwargs.get("force") is True
+    # #98206: the live session row id must also ride along as task_id — the
+    # main gateway turn scopes tool dedup state (skill_view / read_file) to
+    # session_entry.session_id, so a manual compression that resets under the
+    # "default" fallback leaves the live session's dedup records in place and
+    # a post-compression skill_view returns a stub pointing at pruned context.
+    assert (
+        agent_instance._compress_context.call_args.kwargs.get("task_id") == "sess-1"
+    )
 
 
 @pytest.mark.asyncio
--- a/tests/gateway/test_session_hygiene.py
+++ b/tests/gateway/test_session_hygiene.py
@@ -235,9 +235,15 @@ async def test_session_hygiene_preserves_transcript_when_no_rotation(monkeypatch
             self._print_fn = None
             self.shutdown_memory_provider = MagicMock()
             self.close = MagicMock()
+            self.compress_task_id = None
             type(self).last_instance = self
 
         def _compress_context(self, messages, *_args, **_kwargs):
+            # Capture the task scope the gateway forwards (#98206): the dedup
+            # reset inside compress_context is only effective under the live
+            # session row id, the same task_id the main turn hands to
+            # run_conversation.
+            self.compress_task_id = _kwargs.get("task_id")
             # No session_db → cannot rotate: session_id is UNCHANGED, and this
             # is a failure-to-rotate, not an in-place success.
             return ([{"role": "assistant", "content": "summary only"}], None)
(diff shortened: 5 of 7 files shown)


liyangbing:
<h3>Review</h3><p>This is a focused fix for a concrete scope mismatch: the live turn records tool-read deduplication under the session row id, while the two out-of-turn compression paths previously reset the fallback <code>default</code> scope. Forwarding <code>session_entry.session_id</code> in both <code>gateway/run.py</code> and <code>gateway/slash_commands.py</code> matches the identity described in the PR.</p><h3>Suggested contract</h3><ul><li>Make the session-row-id requirement explicit at the helper boundary, so a future caller cannot silently pass a session key or <code>None</code> and recreate the fallback-scope bug.</li><li>Keep dedup state isolated across two simultaneous sessions: compressing session A must clear A's records without changing an unchanged <code>skill_view</code>/<code>read_file</code> result in session B.</li><li>Cover both outcomes claimed in the PR: in-place compaction must invalidate the live session's records, while rotated compaction must not let an old row's stub affect the new row.</li><li>Exercise a stale tool-read stub after an out-of-turn <code>/compress</code> and after the hygiene sweep, including repeated compression and a tool read immediately before/after the reset.</li><li>Preserve any intentional default-scope behavior for non-gateway callers; the regression should verify that the fix is routing the correct scope, not globally removing the fallback semantics.</li></ul><h3>Regression matrix</h3><ul><li><code>/compress</code> and session-hygiene compression, each with unchanged <code>skill_view</code> and <code>read_file</code> content</li><li>in-place versus rotated compaction, with old and new session row ids</li><li>two concurrent sessions and a repeated reset, proving no cross-session dedup invalidation</li><li>tool read before reset, during the compression boundary, and after reset</li><li>non-gateway/default-scope callers and a missing or malformed session identity</li></ul><p>The reported 34 focused passes and the red-before/green-after assertions are strong evidence. The isolation and rotation cases above would protect the session-identity contract if another compression entry point is added later.</p>

liuhao1024:
Thanks for the thorough review — pushed `59c995c` to address the contract points:

**Boundary made explicit.** `compress_context`'s docstring now states the task_id contract directly: gateway callers must pass the live session row id (`session_entry.session_id`, the same task_id the main turn hands to `run_conversation`); `None` is documented as not a valid scope here because both resets treat it as *clear every task*, which would break concurrent-session isolation rather than recreate the fallback-scope bug; `"default"` remains correct only for non-gateway callers with no per-session scope.

**Isolation and rotation pinned at the reset boundary** (`tests/tools/test_file_read_guards.py`, `tests/tools/test_skill_view_dedup.py`):
- Concurrent sessions: compressing session A — repeated, i.e. `/compress` then the hygiene sweep — clears only A's records in both the file-read and skill_view trackers; session B's unchanged `read_file`/`skill_view` dedup survives.
- Rotation: a rotated session's new row id never inherits the old row's stub (fresh scope reads return full content after the old row's reset).
- The non-gateway fallback stays pinned: a bare `reset_file_dedup()` clears every scope.

Two deliberate scope choices on the matrix:
- For session B's survival I assert the tracker state directly instead of issuing a third `read_file` — a third consecutive stub escalates into the #15759 stub-loop guard, which is a separate contract from scope isolation.
- In-place vs rotated both reach the same post-commit reset point in `compress_context`, so their observable difference is exactly the scope semantics covered above (live row id cleared, new row id fresh); the gateway-side forwarding assertions for both entry points (`/compress` and hygiene) were already in the original diff.

kshitijk4poor:
Salvaged into #115298 (auto-merge armed, rebase) — your fix commit `78f8ece4ec` is cherry-picked there with your authorship preserved (`be11d2a34d` on the branch), resolved onto the decomposed gateway: the generic hygiene sweep now lives in `gateway/run_turn.py`, the codex hygiene in `gateway/run.py`, and the manual `/compress` sites were already fixed by #113194 (`b002dfc04d3c`). Two follow-ups ride on top: the codex manual `/compress` site and an `or "default"` guard so an empty session id cannot reset every task's read dedup.

Your session-row-id vs session-key call was the load-bearing part — credited in the PR body. Thank you. Closing this one in favour of #115298; if the merged result misses anything you covered, say so there and I will reopen.

Issue #35800, opened 2026-05-31
[Feature]: .library protected cold storage for skills + intelligent Librarian + enhanced Curator compactor


apoapostolov:
# [Feature Proposal]: .library — Protected Cold Storage for Skills + Intelligent Librarian + Curator Compactor

**Related issues:**
- #2045 (Lazy skill loading — remove full skill listing from system prompt)
- #18900, #26451, #18809 (skill discovery leaks into .archive and hidden dirs)
- Curator implementation (agent/curator.py and docs)

> **Edit note (2026-05-31):** Original draft contained specific personal skill names. This version has been sanitized for privacy.

## Summary

Introduce a new **protected, hard-excluded directory** `~/.hermes/skills/.library/` (and support for it under external_dirs and shared skill repos).

- Skills placed here are **completely invisible** to the normal skill index that gets injected into every system prompt.
- Hermes gains a "Librarian" behavior that activates when the agent or user refers to a skill that isn't currently available in the active catalog.
- The existing Curator gets a new "compactor" mode that identifies long-dormant skills (across user, shared, and agent-created) and proposes (or executes) moves into .library to reclaim prompt tokens.
- Activation is always explicit user approval (move/copy from .library into a real category folder).

This directly attacks the skill catalog bloat problem that power users (especially those syncing large shared-skills repos with 50-100+ niche skills) experience today.

## The Core Problem (Beyond #2045)

Even with lazy body loading (SKILL.md content only read on `skill_view`), the **skill index** (name + description for every skill) still gets dumped into the prompt on every turn for many users.

Current reality for heavy users:
- Bundled skills + optional skills + external_dirs from shared skill collections (covering categories such as devops, github, lifestyle, creative, integrations with external tools, etc.).
- Dozens of highly specific, occasionally useful skills: hardware and machine maintenance tools, creative/content generation skills, game server and VTT tooling, document/PDF processing variants, social media and service monitors, integration families for third-party applications, and many narrow one-off utilities.
- Many of these are used once every few months or only in specific contexts (cron/no_agent scripts, one-off maintenance, specialized campaign prep).

Result: the `<available_skills>` block keeps growing. Every message on Discord/Telegram pays the tax. The Curator helps a little (only for agent-authored skills, moves to `.archive`), but:
- `.archive` still leaks in some discovery paths.
- No systematic way for the *agent* to know "this skill exists but is parked".
[...]

PR #98216, opened 2026-08-30
fix(skills): hard-exclude .library cold-storage dir and close .archive leak paths

DavidMetcalfe:
## Summary

Hidden cold-storage skill dirs must be invisible to skill discovery. The exclusion mechanism is centralized in `EXCLUDED_SKILL_DIRS` (`agent/skill_utils.py`), but seven scan sites bypassed it with raw `rglob`/`glob` or direct-path joins — so `.archive` (the curator's recoverable archive) leaked through them, and a proposed `.library` cold-storage dir (issue #35800) would have leaked identically.

This diff:

- Adds `.library` to `EXCLUDED_SKILL_DIRS` so the layout proposed in #35800 is hard-excluded from day one.
- Closes the seven bypass sites so hidden storage dirs are excluded everywhere:
  - `tools/skills_tool.py` — `skill_view` resolution Strategies 1/1b/3 now consult `is_excluded_skill_path`. **Bare names never resolve into excluded dirs; explicit path addresses still resolve**, preserving the preview-before-restore workflow for archived/parked skills (`hermes curator restore` unaffected).
  - `agent/learning_graph.py` — rglob walk prunes excluded paths.
  - `tools/blueprints.py` — blueprint lookup ignores excluded dirs.
  - `hermes_cli/web_server.py` — profile skill-disable scan no longer counts archived/library skills as installed (prevents archived skill names from being written into the disabled set).
  - `tools/skills_sync_client.py` — installed-skill and org-mirror listings skip excluded storage contents.
- Leaves `tools/credential_files.py` unchanged intentionally: it is a security scanner enumerating symlinks/credentials — scanning excluded dirs is desired there.

## Design note: bare names vs explicit paths

The one behavior change: `skill_view("archived-skill")` for a skill parked under `.archive/`/`.library/` no longer resolves when a legacy flat `<name>.md` match is the only candidate, and bare excluded dir names (`.archive`, `.library`) are not addressable. Explicit relative paths (`.archive/2026-08/old-skill`) still resolve, so an agent or user can preview a parked skill's `SKILL.md` before deciding to restore it. This split was flagged as a required condition in cross-vendor review of the approach (both reviewers conditioned the fix on preserving preview-before-restore).

## Testing

- New `tests/tools/test_skill_exclusion_hardening.py` (12 tests): RED-first (6 failed pre-fix, exactly the leak paths), all pass post-fix. Covers frozenset membership, index invisibility, bare-name resolution refusal (including bare `.archive`/`.library`), flat-md leaks, explicit-path preview, traversal inside excluded dirs, learning-graph and blueprints scans.
- Targeted regression: `test_skill_view_traversal`, `test_skill_view_path_check`, `test_skill_view_dedup`, `test_skill_utils`, `test_blueprints`, `test_learning_graph`(+render), `test_skills_sync`, `test_skills_sync_client`, `test_skills_tool`, `test_curator_archive_prune`, `test_curator_run` — 158 passed.

## Notes

The same hardening closes the `.archive` leak paths that issue #35800 identified ("`.archive` still leaks in some discovery paths"), so the librarian/cold-storage proposal there has a correct foundation. The librarian/compactor design discussion itself is larger than this diff and tracked on the issue.

Open question: should `skill_view`'s explicit-path preview inside excluded dirs eventually require an opt-in flag rather than being always available? Deferred as a UX decision — current behavior preserves the existing curator workflow.

### DIFF
--- a/agent/learning_graph.py
+++ b/agent/learning_graph.py
@@ -75,9 +75,13 @@ def _category(fm: dict[str, Any], skill_md: Path) -> str:
 
 
 def _iter_skill_files(roots: list[tuple[str, Path]]):
+    from agent.skill_utils import is_excluded_skill_path
+
     for source, root in roots:
         if root.exists():
             for path in root.rglob("SKILL.md"):
+                if is_excluded_skill_path(path):
+                    continue
                 yield source, path
 
 
--- a/agent/skill_utils.py
+++ b/agent/skill_utils.py
@@ -1233,6 +1233,16 @@ def iter_skill_index_files(skills_dir: Path, filename: str):
             if d not in EXCLUDED_SKILL_DIRS
             and not (has_skill_md and d in SKILL_SUPPORT_DIRS)
         ]
+        # Symlink-aware exclusion: os.walk(followlinks=True) yields the
+        # symlink's lexical path, not the target's. A symlink whose target
+        # lives inside .archive/.library passes the lexical filter above.
+        # Check the resolved root to close this leak.
+        try:
+            resolved = Path(root).resolve()
+            if any(part in EXCLUDED_SKILL_DIRS for part in resolved.parts):
+                continue
+        except OSError:
+            pass
         if filename in files:
             matches.append(os.path.join(root, filename))
     for path in sorted(matches):
--- a/agent/skill_utils.py
+++ b/agent/skill_utils.py
@@ -31,6 +31,7 @@ EXCLUDED_SKILL_DIRS = frozenset(
         ".github",
         ".hub",
         ".archive",
+        ".library",
         ".venv",
         "venv",
         "node_modules",
--- a/hermes_cli/web_server.py
+++ b/hermes_cli/web_server.py
@@ -15196,7 +15196,11 @@ def _disable_unselected_skills(profile_dir: Path, keep: List[str]) -> int:
         installed: List[str] = []
         skills_root = profile_dir / "skills"
         if skills_root.is_dir():
+            from agent.skill_utils import is_excluded_skill_path
+
             for md in skills_root.rglob("SKILL.md"):
+                if is_excluded_skill_path(md):
+                    continue  # .archive/.library/... contents are not installed skills
                 installed.append(md.parent.name)
         cfg = load_config()
         disabled = get_disabled_skills(cfg)
--- a/tests/tools/test_skill_exclusion_hardening.py
+++ b/tests/tools/test_skill_exclusion_hardening.py
@@ -104,6 +104,28 @@ class TestSkillViewExcludedByName:
         )
         assert result["success"] is False
 
+    def test_symlink_into_excluded_dir_not_indexed(self, fake_skills, tmp_path):
+        """A symlink whose target lives inside an excluded dir is not indexed.
+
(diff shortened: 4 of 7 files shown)


liyangbing:
<h3>Review</h3><p>This patch closes the documented <code>.archive</code>/<code>.library</code> discovery paths across skill indexing, learning-graph scans, blueprint lookup, profile disablement, and sync listings. The explicit-path preview contract is also tested, which is important for restore workflows.</p><h3>Suggested contract</h3><ul><li>Keep the exclusion rule centralized and apply it to every filesystem walk, including future scan sites; add a small invariant test that enumerates the supported discovery helpers.</li><li>Test path identity at the boundary: symlinks whose lexical path is excluded, symlinks that resolve into an excluded directory, Windows case/separator variants, and traversal attempts from an explicit preview path.</li><li>Clarify whether explicit preview is intentionally allowed for every excluded directory or only curator-owned storage. If it remains allowed, ensure read, metadata, and any companion-file access use the same containment check.</li><li>Exercise the race where a skill is moved into <code>.archive</code> or <code>.library</code> after discovery but before read/sync/disablement, and confirm the final canonical-path check still excludes it.</li></ul><h3>Regression matrix</h3><ul><li>active, archived, and library skills through index, <code>skill_view</code>, learning graph, blueprints, sync, and profile-disable scans</li><li>bare-name refusal versus explicit-path preview, including flat legacy Markdown files</li><li>symlink and traversal cases, mixed path spellings on Windows, and a move during preflight</li><li>an excluded skill with the same name as an active skill, plus a failed or partial sync</li></ul><p>The reported 158 targeted passes are useful evidence; the boundary and race cases above would make the hard-exclusion guarantee easier to preserve as new discovery paths are added.</p>

DavidMetcalfe:
@liyangbing — thanks for the review. Responses to each suggestion:

**Symlink handling** — good catch. `os.walk(followlinks=True)` yields the symlink's lexical path, not the target's, so a symlink into `.archive/`/`.library/` passed the lexical `EXCLUDED_SKILL_DIRS` filter. Fixed in `26bcb21` by adding a resolved-root check in `iter_skill_index_files` (same pattern the sync client already uses at `skills_sync_client.py:536`). Regression test added: symlink-into-excluded-dir is now not indexed.

**Name-collision test** — accepted. Added `test_excluded_name_shadowing_active_skill`: an excluded skill with the same name as an active skill does not shadow it (the active skill resolves, the excluded one doesn't).

**Explicit preview scope** — already addressed. The PR's design note explains the contract (bare names never resolve into excluded dirs; explicit paths preview for restore workflows). `skill_view` has a containment check at `skills_tool.py:1547` ensuring resolved paths stay within the skill directory. The question of whether preview should eventually require an opt-in flag is deferred as a UX decision in the PR notes.

**Remaining suggestions** — the invariant test enumerating discovery helpers and the race-condition test are valid observations but speculative (testing for code that doesn't exist yet or pre-existing TOCTOU across all skill operations). Windows path variants are irrelevant for this project (macOS/Linux). The regression matrix rows for active/archived/library through all scan paths, bare-name vs explicit-path, and flat legacy markdown are already covered by the existing 14 tests.

PR #98217, opened 2026-08-30
fix(desktop): strip custom: prefix from provider label in UI

VALIKO26:
## Problem

When a custom provider is configured under the `providers:` key in ~/.hermes/config.yaml (e.g. `providers: yjs_github_artur: {...}`), the backend canonicalises the identity to `custom:<config-key>` (see `hermes_cli/providers.py::canonical_custom_identity` and `custom_provider_aliases`). The Desktop UI then renders this raw string verbatim:

- The composer model-pill tooltip showed `Model · custom: <model-name>`
- The model-picker dialog header showed `Current: <model> · custom`

In the `model.options` catalog the same providers correctly display their config-key as the human-facing name (slug=`yjs_github_artur`, name=`yjs_github_artur`), so showing `custom` to the user is both inconsistent and uninformative — the user cannot tell \*which\* of their custom providers is active.

A second, related path `state.provider === 'custom'` (no key suffix) is documented in `hermes_cli/providers.py` as a known '*corrupt state from a prior model-switch bug*' and is already normalised / fallback-healed at the backend. The UI should never surface it.

## Fix

Add a tiny pure helper `displayProviderName()` next to the existing `displayModelName()` in `apps/desktop/src/lib/model-status-label.ts`:

- `custom:<key>` → `<key>` (the config key IS the user-facing name)
- bare `custom` → empty string (corrupt: don't surface a meaningless label, let the caller fall back to its no-provider rendering)
- anything else passes through unchanged

Wire it at the two places that render the current provider for the user:

1. `apps/desktop/src/app/chat/composer/model-pill.tsx` — the tooltip on the model pill next to the input.
2. `apps/desktop/src/components/model-picker.tsx` — the `Current: …` dialog description.

## Tests

4 new cases in `apps/desktop/src/lib/model-status-label.test.ts` cover:
- `custom:<key>` strips the prefix
- bare `custom` drops to empty
- empty / whitespace input stays empty
- regular slugs pass through unchanged (incl. `custom-foo` which has no colon)

```
$ npx vitest run src/lib/model-status-label.test.ts
Test Files  1 passed (1)
     Tests  15 passed (15)
```

## Notes

- No backend change required: gateway continues to send the canonical `custom:<key>` identity (necessary for aliasing in `model.options`), the UI just formats it.
- The fix is a pure display layer — it does not affect model selection, provider detection, or session state persistence.
- Follows the existing `displayModelName()` pattern in the same file, mirroring the conventions established there.

### DIFF
--- a/apps/desktop/src/app/chat/composer/model-pill.tsx
+++ b/apps/desktop/src/app/chat/composer/model-pill.tsx
@@ -11,7 +11,7 @@ import { releaseTypingFocus } from '@/components/ui/keyboard-first'
 import { Tip } from '@/components/ui/tooltip'
 import { useI18n } from '@/i18n'
 import { ChevronDown } from '@/lib/icons'
-import { formatModelStatusLabel } from '@/lib/model-status-label'
+import { formatModelStatusLabel, displayProviderName } from '@/lib/model-status-label'
 import { cn } from '@/lib/utils'
 import { $currentModelSource, $defaultReasoningEffort, setModelPickerOpen } from '@/store/session'
 
@@ -127,8 +127,9 @@ export function ModelPill({
       )
     : PILL
 
-  const baseTitle = currentProvider
-    ? copy.modelTitle(currentProvider, currentModel || copy.modelNone)
+  const providerLabel = displayProviderName(currentProvider)
+  const baseTitle = providerLabel
+    ? copy.modelTitle(providerLabel, currentModel || copy.modelNone)
     : copy.switchModel
 
   const title = pinnedOverride ? `${baseTitle} — ${copy.modelPinned}` : baseTitle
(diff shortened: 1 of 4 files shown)


Enough1122:

**Overall:** Clean UI label normalizer for `custom:<key>` canonical identities. Prevents raw `custom:` protocol token leaking into composer pill and picker header.

**What it does**
- `apps/desktop/src/lib/model-status-label.ts:117` adds `displayProviderName(provider)` — trims, lower-checks `custom` → `""`, `custom:` → `raw.slice(7).trim()`, else returns `raw` unchanged. Documents that bare `custom` is corrupt state from known backend bug and should fall back to no-provider wording.
- Wires into `model-pill.tsx:20` (`copy.modelTitle(providerLabel, ...)`) and `model-picker.tsx:44` (header `Current: ... · ...`).
- Adds `model-status-label.test.ts:67` covering `custom:key` stripping, bare `custom` collapse, canonical slug passthrough, empty/whitespace.

**Non-blocking notes**
- `lower.startsWith('custom:')` but `raw.slice(7)` preserves original case of the key portion — intentional (config key case preserved). Confirm keys are case-sensitive downstream (they are).
- Empty result for bare `custom` means pill falls back to `switchModel` / picker omits `·` suffix — desired per spec; double-check a11y label still meaningful when provider collapses.
- Safe for already-canonical slugs (`openai-codex`, etc.) — returned verbatim.

PR #95281, opened 2026-08-26
pm: unified package manager

ethernet8023:
Everything hermes depends on is a package in one dependency tree, split across three files with one authority each:

- package definitions (pm/packages.py): url templates, layout, deps, staging, declared platform gaps. Humans write these.
- lock.json (committed): exact versions + sha256 per target, written by `pm lock --bump`, never by hand. Plugins ship the same document kind.
- facts.json (per install): what is actually on this machine — entry, version, target, artifact digests, env, stamp. Written only by pm.

Rules: package managers (uv, npm) are pm internals — no code outside pm/ runs them or holds their paths; consumers name outcomes (pm.run("ffmpeg"), sync_venv(["telegram"])). Installing happens at install time; startup is one O(1) stamp check that never touches the network.

The venv is one package: pyproject + uv.lock + enabled extras, made true by uv sync inside pm. Feature names ARE extra names — the 42-name LAZY_DEPS mapping dies. tools/lazy_deps.py deleted (~1,900 lines with its machinery tests); ~35 call sites now use pm.ensure_import/pm.available. hermes update's dependency phase collapses to pm.sync_venv (uv/pip ladders, lazy-refresh repair machinery, editable-install currency probe all deleted). Termux install path ripped per standing policy: extras, installer arm, psutil-android prebuild, constraints file, bootstrap tiers.

Downloads are store entries keyed by hash; the store is machine-wide, immutable, gc'd by reachability (with a grace window so in-flight downloads survive a sweep). Startup check wired into main() before dispatch. hermes pm CLI: lock/install/env/doctor/gc.

## What does this PR do?

This PR replaces hermes-agent's ~6 overlapping dependency mechanisms (lazy-deps, requirements ladders, runtime-pins, per-plugin node probing, installer-side tool staging) with **one package manager under `pm/`**, plus the reviewer-driven hardening and cleanup pass on top of the foundation:

**Install-state authority (reviewer P1 #1/#3).** facts.json now records the full identity of every installed package: `target`, artifact `sha256` list, and a `digest` of the realized bytes (a deterministic `tree_digest()` over an ordered, symlink-aware walk). `installed()` compares recorded identity against the current lock — a same-version re-pin, a stale entry, or a corrupted cache can no longer masquerade as installed. The fetch cache re-hashes cached downloads against the full digest before reuse. `pm doctor` re-hashes realized bytes against the recorded digest and flags legacy identity-less facts. Re-realizing an established package logs a `repair:` line.

[...]

PR #98223, opened 2026-08-30
pm clean

ethernet8023:


alt-glitch:

Duplicate of #95281: this is the same broad unified package-manager implementation, with only a small pm/cli.py cleanup difference. #95281 is the earlier open review thread.

PR #98226, opened 2026-08-30
feat(desktop): model picker details panel, dedicated reasoning pill & Free tier badge

henrickgu:
## Summary

Enrich the Desktop model picker with the model facts users need to choose consciously, modeled on OpenCode's per-model tooltip, and split reasoning effort into its own composer control.

- **Per-model details panel on hover** (OpenCode-style label/value rows): Model, Provider, **Inputs** (text/image/audio/video/pdf from `input_modalities`), Context window, Max output, and pricing when the provider reports it.
- **Per-row summary** (`163.8K · Vision · Tools`) so facts are visible without opening the panel.
- **Free tier badge** derived from the models.dev record/`-free` id suffix (e.g. `muse-spark-1.2-contributor-free`) — pure display signal, no second catalog, no manual curation.
- **Dedicated composer ReasoningPill** for the active model's effort, inheriting the profile default when the session has none; `ModelPill` keeps fast mode visible (param or `-fast` variant).
- **Metadata projection**: `hermes_cli/inventory.py` projects the existing models.dev registry into `model.options` as an optional `metadata` map (context, output, modalities, capability booleans). Unknown models are omitted — no parallel catalog.
- **Preset key fix**: reasoning preset writes key on the provider row slug, re-resolved from the query cache at click time, so aliased providers (`custom:<key>`) restore their presets.
- Panel labels localized (en/zh; other locales inherit via `defineLocale`).

## Test plan

- [x] `vitest` — 51/51 green across all touched test files (incl. new `ReasoningPill` profile-default test and Free-derivation test)
- [x] `vitest run --changed HEAD` — 4,569 passed; the 9 failures are pre-existing on non-en-US machines (`toLocaleString()` without an explicit locale in `format.ts`/billing/messaging tests — unrelated to this change)
- [x] `tsc -p . --noEmit` clean; `eslint` clean on all touched files
- [x] `pytest tests/hermes_cli -k inventory` — 56 passed (2 pre-existing Windows-only collection failures unrelated to this change: `os.geteuid` and `wmic`-based scan test)
- [x] Live Electron probe: hover flow shows the details panel against a running app

## Notes for reviewers

- The metadata map is additive and optional; consumers that don't read it are unaffected.
- Known follow-up (not in this PR): `opencode-free` is not yet in `PROVIDER_TO_MODELS_DEV` — only matters if the free tier needs picker metadata; `opencode-zen`/`opencode-go` are mapped.


henrickgu:
Closing in favor of #105076, now rebased on current main and scoped to the model metadata panel + Free tier badge (the dedicated reasoning pill was upstreamed via #109907).

PR #41368, opened 2026-06-07
fix(tools): keep compound-background rewrite valid when a command follows

oryn-oryn:
`_rewrite_compound_background` turns `A && B &` into `A && { B & }` to
avoid the subshell-wait process leak. But when a command follows the
backgrounded job on the same line (`A && B & C`), the rewrite produced
`A && { B & } C` — a bash syntax error, since a brace group needs a
separator before the next command. Valid commands like
`build && python app.py & echo started` were rejected with
"syntax error near unexpected token". Splice in a `;` when the tail is a
same-line command; tails that already start with a separator (newline,
`&&`/`||`/`|`) are untouched.

## What does this PR do?

Fixes `_rewrite_compound_background` emitting invalid bash when a command
trails the backgrounded job on the same line.

## Related Issue

N/A

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `tools/terminal_tool.py`: in `_rewrite_compound_background`, insert `;`
  after the brace group when a same-line command follows; leave tails
  beginning with a separator unchanged.
- `tests/tools/test_terminal_compound_background.py`: add regression tests
  for trailing same-line commands, `bash -n` validity, and idempotence.

## How to Test

1. `from tools.terminal_tool import _rewrite_compound_background as r`
2. `r("A && B & C")` → `"A && { B & }; C"` (was `"A && { B & } C"`).
3. `bash -n -c "$(echo a && sleep 100 & echo done)"` style output now
   passes; pre-fix it failed.
4. `pytest tests/tools/test_terminal_compound_background.py -q` → 39 passed.

### Code

- [x] I've read the Contributing Guide
- [x] My commit messages follow Conventional Commits
- [x] I searched for existing PRs to make sure this isn't a duplicate
- [x] My PR contains only changes related to this fix
- [x] I've run the affected tests and they pass
- [x] I've added tests for my changes
- [x] I've tested on my platform: macOS 15 (Darwin 25.5)

### Documentation & Housekeeping

- [x] I've updated relevant documentation — N/A
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — N/A
- [x] I've updated `CONTRIBUTING.md` or `AGENTS.md` — N/A
- [x] I've considered cross-platform impact — pure string logic, platform-agnostic
- [x] I've updated tool descriptions/schemas — N/A

PR #98227, opened 2026-08-30
fix(tools): keep compound-background rewrites parseable when a statement follows the &

### LINKED ISSUE #98222: [Bug]: execute_code remote kernel spawn always fails on Docker/SSH/Modal — _rewrite_compound_background corrupts commands with a statement after &
## Bug Description

Every `execute_code` call on a Docker sandbox (and, same code path, SSH/Modal) fails to spawn the persistent remote kernel. The command issued by the spawn template is valid bash; the wrapper that corrupts it is added by `_rewrite_compound_background` before execution:

```
bash: eval: line 5: syntax error near unexpected token `echo'
bash: eval: line 5: `cd /tmp/hermes_rkernel_04dcca077e03 && { nohup env ... python3 kernel_runner.py > /tmp/hermes_rkernel_04dcca077e03/runner.log 2>&1 & } echo PID:$!'
```

Result: no PID is returned, the kernel spawn gives up, and every call falls open to the per-call path — no variable/import persistence across calls, one interpreter startup per call, and the warning repeats on **every** `execute_code` call.

Note the blast radius is broader than kernel spawn: the rewrite is applied by default to **all** `Environment.execute()` calls on docker/ssh/modal backends (`tools/environments/base.py:1496-1499`), so *any* user command of the form `A && B & C` (a statement following the `&`) is corrupted in those environments.

## Steps to Reproduce

1. Any profile whose sandbox backend is Docker (or SSH/Modal) — e.g. a profile with `execute_code` running in a Docker sandbox.
2. Trigger any `execute_code` call.
3. `tools/code_kernel_remote.py:259-261` runs this template through `env.execute()`:
   ```
   cd /tmp/hermes_rkernel_XXXX && nohup env ... python3 kernel_runner.py > /tmp/hermes_rkernel_XXXX/runner.log 2>&1 & echo PID:$!
   ```
   (valid bash — `&` is itself a statement terminator)
4. `env.execute()` applies `_rewrite_compound_background` by default, producing:
   ```
   cd /tmp/hermes_rkernel_XXXX && { nohup env ... python3 kernel_runner.py > .../runner.log 2>&1 & } echo PID:$!
   ```
5. `bash -n` fails with `syntax error near unexpected token 'echo'` → no PID → spawn returns `None` → `WARNING tools.code_kernel_remote: remote kernel spawn returned no PID`.

Minimal unit repro (no sandbox required):

```python
from tools.terminal_tool import _rewrite_compound_background as rw
out = rw("cd /tmp && nohup server.py > log 2>&1 & echo PID:$!")
print(out)   # cd /tmp && { nohup server.py > log 2>&1 & } echo PID:$!
# bash -n -c "$out"  →  syntax error near unexpected token `echo'
```

## Expected Behavior

The rewrite keeps the command valid. `A && B & C` is well-formed bash today, so the rewritten form must be too: `echo PID:$!` should still run and return the backgrounded runner's PID (the spawn template depends on it to find the kernel).

## Actual Behavior

```
bash: -c: line 1: syntax error near unexpected token `echo'
bash: -c: line 1: `cd /tmp/x && { nohup env A=B python3 k.py > /tmp/x/runner.log 2>&1 & } echo PID:$!'
```

Gateway log (affected profile, 8 repetitions within one minute during a normal session):

```
2026-08-30 03:00:16,623 WARNING tools.code_kernel_remote: remote kernel spawn returned no PID: "bash: eval: line 5: syntax error near unexpected token `echo'..."
```

## Affected Component

- N/A (CLI only)

Tool-layer bug. It surfaced through a Google Chat gateway session, but it reproduces identically in CLI with a Docker sandbox; no platform adapter is involved.

## Debug Report

```
agent.log   https://dpaste.com/6A2922YAL
gateway.log https://dpaste.com/4UYCSABTX
gui.log     https://dpaste.com/9URUTM78N
```

Note: the main `hermes debug share` report was generated, reviewed, and then **deleted** — it contained the live diagnostic session's own log lines (PII). The three log tails above were checked line-by-line and contain no PII or credentials (redaction runs with `force=True`).

- Hermes version: `0.20.6 (2026.8.27)`, git `3340bbbd`, upstream `4209d371`
- OS: Ubuntu 26.04.1 LTS (x86_64)
- Python: 3.11.15 (venv)

See "Actual Behavior" — the warning above repeats once per `execute_code` call on the affected profile.

## Root Cause Analysis

Two shipped commits collide:

1. **The rewrite** — `_rewrite_compound_background`, `tools/terminal_tool.py:860`, introduced in `abfc1847b7` (2026-04-19, "rewrite `A && B &` to `A && { B & }` to prevent subshell leak"). It wraps the backgrounded command after a depth-0 `&&`/`||` in a brace group:
   ```python
   result = prefix + "{ " + middle + "& }" + suffix   # terminal_tool.py:1020
   ```
   The apply-site comment (`terminal_tool.py:1018-1019`) contains the false assumption:
   ```
   # `{` needs a trailing space in bash; the closing `}` needs to be
   # preceded by `;` or `&` — we're providing `&` from the backgrounding.
   ```
   That is only true when the `&` is the **last** token. In `A && B & C` the `&` ends up *inside* the brace group, where it terminates the last command *in* the group — it does not terminate the group. In bash, `}` is a compound-command **delimiter**, not a terminator: a command immediately after `}` without `;`/`&`/newline is a syntax error.

2. **The victim** — the kernel spawn template, `tools/code_kernel_remote.py:259-261`, introduced in `5f75ec197b` (2026-08-28, "remote kernel host — session persistence for docker/ssh/modal backends"). It is the first shipped command with a statement after the `&`, and it postdates the rewrite by ~4 months, so the two never coexisted before this.

Existing tests (`tests/tools/test_terminal_compound_background.py`, 15 tests) never cover the statement-after-`&` form — `test_multiline_idempotent` uses a *newline* after the `&`, which is the one terminator the rewrite happens to handle.

## Proposed Fix

In the apply loop, insert a statement terminator when the suffix begins with a non-whitespace character (`tools/terminal_tool.py:1020`):

```python
suffix = result[amp_pos + 1 :]
if suffix and not suffix[0].isspace():
    suffix = "; " + suffix
result = prefix + "{ " + middle + "& }" + suffix
```

For the kernel template this yields `cd DIR && { nohup ... & }; echo PID:$!` — valid bash, and `echo PID:$!` still runs in the current shell with the correct `$!` (brace groups do not fork), so the template's semantics are preserved exactly. A second pass through the rewrite is idempotent: the in-group `&` is skipped at `brace_depth > 0` and the `;` resets the chain state.

Suggested regression test:

```python
def test_background_followed_by_command(self):
    out = rewrite("cd /tmp && server & echo $!")
    assert "};" in out or "}; " in out
    # and must be valid bash:
    assert subprocess.run(["bash", "-n", "-c", out]).returncode == 0
```

liuhao1024:
## What does this PR do?

`_rewrite_compound_background` rewrote `A && B &` into `A && { B & }` to avoid the subshell-wait process leak, but when a statement followed the `&` on the same line the output was `A && { B & } echo PID:$!` — a hard bash syntax error (`syntax error near unexpected token 'echo'`), because a brace group needs a list terminator (`;`, `&`, or a newline) before the next command.

The remote-kernel spawn template used by `execute_code` on Docker/SSH/Modal backends is exactly this shape (`cd … && nohup … & echo PID:$!`), so the rewritten command failed on every call: no PID came back, the persistent kernel never spawned, and every `execute_code` call silently fell back to the per-call interpreter path (no variable/import persistence, one interpreter startup per call, plus a repeated warning). The corruption applied to **any** user command of the form `A && B & C` on those backends, since the rewrite is the default for `Environment.execute()`.

The fix injects `; ` after the closing `}` only when a plain command follows the `&`. Newlines, `;`, `&&`/`||`, `|`, and `#` comments already separate statements (or would be double-terminated), so they pass through untouched. `$!` still points at the backgrounded command, preserving the `echo PID:$!` contract.

## Related Issue

Fixes #98222

## Type of Change

## Changes Made

- `tools/terminal_tool.py` — `_rewrite_compound_background`: when the text after the backgrounding `&` starts a new plain command, splice `; ` in place of the intervening whitespace so the brace group is properly terminated; tail shapes that already separate statements (`\n`, `;`, `&`, `|`, `#`) keep the old pass-through.
- `tests/tools/test_terminal_compound_background.py` — new `TestStatementAfterBackground`: the #98222 kernel-spawn shape, a minimal `A && B & echo done`, newline/comment tails staying unmodified, and idempotence of the rewrite with a tail.

## How to Test

1. `pytest tests/tools/test_terminal_compound_background.py -q` — 20 passed (15 pre-existing + 5 new), all pre-existing expected outputs unchanged.
2. `pytest tests/tools/test_terminal_tool.py -q` — 10 passed (neighbor rewriter tests unaffected).
3. End-to-end proof of the reported failure:
   - Pre-fix shape `cd /tmp && { nohup sleep 5 >/dev/null 2>&1 & } echo PID:$!` under `bash -c` exits 2 with `syntax error near unexpected token 'echo'` — the exact error from the issue.
   - Post-fix rewrite of `cd /tmp && nohup sleep 5 >/dev/null 2>&1 & echo PID:$!` produces `cd /tmp && { nohup sleep 5 >/dev/null 2>&1 & }; echo PID:$!`, which exits 0, prints `PID:<pid>`, and `ps -p <pid>` confirms the backgrounded `sleep` is alive.

### Code

- [x] I've tested on my platform: macOS 15 (arm64), bash 3.2/5.x

### DIFF
--- a/tests/tools/test_terminal_compound_background.py
+++ b/tests/tools/test_terminal_compound_background.py
@@ -100,3 +100,44 @@ class TestEdgeCases:
 
     def test_tabs_between_tokens(self):
         assert rewrite("A\t&&\tB\t&") == "A\t&&\t{ B\t& }"
+
+
+class TestStatementAfterBackground:
+    """A statement after the `&` needs a separator after the closing `}`.
+
+    bash requires a list terminator before the next command: ``{ B & } echo``
+    is a hard syntax error (``syntax error near unexpected token 'echo'``),
+    which is exactly how the #98222 kernel-spawn template died — the
+    rewritten ``cd … && nohup … & echo PID:$!`` never returned a PID, so
+    every execute_code call on docker/ssh/modal fell off the persistent
+    kernel. ``{ B & }; echo`` parses and keeps ``$!`` pointing at B.
+    """
+
+    def test_kernel_spawn_shape_gains_separator(self):
+        cmd = (
+            "cd /tmp/hermes_rkernel_04dcca && "
+            "nohup env -i python3 kernel_runner.py > runner.log 2>&1 & "
+            "echo PID:$!"
+        )
+        assert rewrite(cmd) == (
+            "cd /tmp/hermes_rkernel_04dcca && "
+            "{ nohup env -i python3 kernel_runner.py > runner.log 2>&1 & }; "
+            "echo PID:$!"
+        )
+
+
+    def test_simple_command_tail_gains_separator(self):
+        assert rewrite("A && B & echo done") == "A && { B & }; echo done"
+
+
+    def test_newline_tail_already_separates(self):
+        assert rewrite("A && B &\nC") == "A && { B & }\nC"
+
+
+    def test_comment_tail_already_separates(self):
+        assert rewrite("A && B & # note") == "A && { B & } # note"
+
+
+    def test_rewrite_with_tail_stays_idempotent(self):
+        once = rewrite("A && B & echo PID:$!")
+        assert rewrite(once) == once
--- a/tools/terminal_tool.py
+++ b/tools/terminal_tool.py
@@ -1017,7 +1017,20 @@ def _rewrite_compound_background(command: str) -> str:
         suffix = result[amp_pos + 1 :]
         # `{` needs a trailing space in bash; the closing `}` needs to be
         # preceded by `;` or `&` — we're providing `&` from the backgrounding.
-        result = prefix + "{ " + middle + "& }" + suffix
+        # A statement following the `&` on the same line additionally needs a
+        # list separator after the `}`: bash rejects `{ B & } echo` outright
+        # while `{ B & }; echo` keeps `$!` pointing at the backgrounded B
+        # (#98222 — the kernel-spawn `cd … && nohup … & echo PID:$!` template
+        # died on this exact syntax error). Newlines, `;`, `&&`/`||`, pipes,
+        # and comments already separate statements, so only plain commands
+        # get the injected `;`.
+        separator_needed = False
+        tail = suffix.lstrip(" \t")
+        if tail and not tail.startswith(("\n", ";", "&", "|", "#")):
+            separator_needed = True
+        result = prefix + "{ " + middle + "& }" + (
+            "; " + tail if separator_needed else suffix
+        )
 
     return result
 


alt-glitch:

Duplicate of #41368: the diff uses the same semicolon-restoration mechanism in `_rewrite_compound_background` for a same-line command after `&`, with equivalent regression coverage.

liuhao1024:
Fair call — withdrawing in favor of #41368, which predates this PR by ~3 months and lands the same semicolon-splice fix in `_rewrite_compound_background`.

For context: this PR was filed in good faith against the freshly-opened #98222 (the remote-kernel spawn failure) before triage linked it back to the older PR, so the overlap only became visible afterwards. Closing this one.

Two small things from my test pass that may be worth folding into #41368:

- The exact failure shape from #98222 is `cd … && nohup … & echo PID:0` — a regression test using that real kernel-spawn shape directly documents why the fix matters (the rewritten command failed `bash -n`, so no PID ever came back and `execute_code` silently fell off the persistent kernel on docker/ssh/modal).
- A tail starting with `#` (e.g. `A && B & # note`) already separates statements; the current exclusion set `(";", "&", "|", "\n", ")")` splices a redundant `; ` before the comment. Harmless, but skipping `#` tails keeps the rewrite minimal.

liuhao1024:
Closing as a duplicate of #41368 (older PR, equivalent fix). Keeping the branch for reference; the maintainers are welcome to cherry-pick the kernel-spawn regression test from here.

PR #73285, opened 2026-07-28
feat(agent): add native Ollama /api/chat adapter (#4505)

hanzlamateen:
## Description

Adds a native Ollama `/api/chat` adapter so local-Ollama users get **honored per-request `num_ctx`** and **correct tool calling**, neither of which the current OpenAI-compatible `/v1` path delivers.

> **Supersedes #55606.** That PR was auto-closed under the `env-var-for-config` policy because it gated on a `HERMES_OLLAMA_NATIVE` env var. That's fixed here — **the env var is gone**; selection is now probe-only, the same detection-based pattern the in-tree `agent/gemini_native_adapter.py` already uses. I couldn't reopen #55606 (the API refuses once a branch gets commits after close, and its branch was ~5k commits behind `main`), so this is a fresh branch rebased onto current `main` with the same work. @laurinaitis reviewed the original end-to-end, confirmed the design, and dropped their competing transport adapter to consolidate on this one.

Local Ollama is configured as `provider: custom` with `base_url: …:11434/v1`. That `/v1` surface:

1. **Silently drops per-request `num_ctx`.** Hermes already does the right thing on its side — `agent_init` detects the model's context via `query_ollama_num_ctx` (`/api/show`) and the `custom` provider profile injects it as `extra_body.options.num_ctx`. But Ollama's `/v1` maps only a fixed allowlist of OpenAI params and ignores the `options` object, so that value never reaches the model — it loads at the **4096** default anyway. The detection/injection machinery is effectively inert over `/v1`; native `/api/chat` is the endpoint that honors `options.num_ctx`. Verified live on Ollama 0.22.0 **and 0.32.5 (latest)**, and confirmed in the v0.32.5 `/v1` handler source (it maps only a fixed OpenAI allowlist to model options — no `num_ctx`); tracked in ollama/ollama#11409 (still open).
2. **Can corrupt streamed `tool_calls` deltas** for local models — the recurring complaint in #4505, and untouched by the num_ctx work. Several users confirm the *same* model returns correct tool calls via native `/api/chat` but broken ones via `/v1`, which leaves an agentic assistant unable to act locally.

This PR adds `agent/ollama_native_adapter.py` — a drop-in, OpenAI-SDK-shaped client over Ollama's native `/api/chat`, **modeled directly on the existing `agent/gemini_native_adapter.py`** so `api_mode` stays `chat_completions` and the rest of the agent loop is unchanged.

### Selection is detection-based, exactly like the Gemini adapter — no feature flag

[...]

Issue #43900, opened 2026-06-11
[Bug] Ollama local models silently capped at 4096-token context — causes finish_reason=length and garbled retry responses


jhonymiler:
## Summary

When using a local Ollama instance with Hermes Agent, the model runs with Ollama's **default `num_ctx` (4096 tokens)** even though the GGUF metadata reports a much larger context (e.g. 131,072 for Gemma 4). Hermes reads the GGUF value and stores it in `_ollama_num_ctx`, but never actually sends it to Ollama's `/v1/chat/completions` endpoint — because the OpenAI-compat route **ignores** the `extra_body.options.num_ctx` field that the `custom` provider profile injects.

The result: once the conversation grows (many turns + tool definitions), the input token count exceeds 4096 and every response returns `finish_reason="length"` with empty or truncated content — even when the Hermes UI shows `0 / 131.1K` context usage. The truncation-retry loop then concatenates three broken partial responses into garbled output.

HaozheZhang6:
Reproduced this against Ollama v0.30.7 — `ollama ps` reports the loaded context window, which is a clean signal for whether `num_ctx` actually took effect:

| request | endpoint | loaded context (`ollama ps`) |
|---|---|---|
| `options.num_ctx: 8192` (what Hermes sends today) | `/v1/chat/completions` | 4096 |
| top-level `num_ctx: 8192` | `/v1/chat/completions` | 4096 |
| `options.num_ctx: 8192` | native `/api/chat` | **8192** |

So `extra_body.options.num_ctx` is silently dropped by the OpenAI-compat route, and there's no `/v1` form that works — top-level `num_ctx` is ignored too (the PR that would have added it, ollama/ollama#6137, was closed unmerged). Pre-loading the model at 8192 via `/api/chat` first doesn't help either: the next `/v1` call reloads it back to 4096, because `/v1` always applies Ollama's server default (4096, or `OLLAMA_CONTEXT_LENGTH`).

Net: `_ollama_num_ctx` can't be honored through the OpenAI-compat path at all. The only ways to raise it are the native `/api/chat` endpoint, a Modelfile `PARAMETER num_ctx`, or the `OLLAMA_CONTEXT_LENGTH` env var. Also worth noting that `test_ollama_num_ctx` only asserts the value lands in `extra_body.options.num_ctx`, so it stays green even though the value never reaches the model.

[...]

PR #98229, opened 2026-08-30
fix(provider): preserve configured num_ctx for local Ollama

wang0090:
## Summary

Fix local Ollama context truncation when Hermes is configured with `ollama_num_ctx`.

Hermes was correctly preserving conversation history and passing `extra_body.options.num_ctx`, but Ollama's OpenAI-compatible `/v1/chat/completions` path did not honor the configured context size in the tested local setup. This caused long multi-turn prompts to be truncated even though history was present in the request payload.

For confirmed local Ollama endpoints with `ollama_num_ctx` configured, this change routes chat requests through Ollama's native `/api/chat` transport so the configured `options.num_ctx` is actually applied.

Other providers and remote OpenAI-compatible endpoints remain unchanged.

## Changes

- add a local Ollama native chat adapter
- gate native routing to confirmed local Ollama endpoints
- preserve configured `num_ctx`
- support both streaming and non-streaming paths
- preserve tool-call and reasoning mappings
- add metadata-only message-shape observability
- add regression coverage for conversation history/provider payload behavior

## Validation

Targeted regression:
- two-turn continuity: PASS
- persistence reload with new agent/DB handle: PASS
- compressor off-path: PASS
- sanitizer safety: PASS
- native Ollama payload preserves history + `num_ctx`: PASS
- result: 5/5 PASS

Live local Ollama (`qwen3.5:9b`):
- Round 1 history=0
- Round 2 history=2
- Round 2 prompt tokens > Round 1
- prior-turn values recalled correctly
- native `/api/chat` confirmed
- `num_ctx=65536` confirmed

Streaming + tool-call smoke test:
- tool schema send: PASS
- tool call receive: PASS
- tool result continuation: PASS
- history preservation: PASS
- streaming: PASS
- no repeated tool call / schema crash / provider fallback

Static validation:
- `py_compile`: PASS
- `git diff --check`: PASS

## Scope

The native transport path is limited to:
- local endpoint
- endpoint confirmed to be Ollama
- `ollama_num_ctx` configured
- chat-completions API mode

Ollama Cloud, OpenAI, vLLM, llama.cpp, and remote OpenAI-compatible providers are not routed through this adapter.

## Notes

The full pytest suite was not run because the existing development environment did not have pytest installed. No packages were installed as part of this fix. The targeted regression tests and live Ollama integration checks above passed.

### DIFF
--- a/agent/chat_completion_helpers.py
+++ b/agent/chat_completion_helpers.py
@@ -933,6 +933,13 @@ def _dispatch_nonstreaming_api_request(agent, api_kwargs: dict, *, make_client):
     interrupt, abort, cancellation, and close semantics stay in the callers —
     this helper only issues the request.
     """
+    from agent.ollama_native_adapter import (
+        create_native_ollama_chat,
+        should_use_native_ollama,
+    )
+
+    if should_use_native_ollama(agent):
+        return create_native_ollama_chat(agent, api_kwargs)
     if agent.api_mode == "codex_responses":
         request_client = make_client("codex_stream_request")
         return agent._run_codex_stream(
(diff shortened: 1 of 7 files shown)


alt-glitch:

Related: #73285 addresses the same native Ollama /api/chat direction through a client facade and broader endpoint probing, while this PR dispatches at call sites for confirmed local Ollama with ollama_num_ctx. Both target the #43900 context-window family; please consider consolidating the approaches.

Issue #68915, opened 2026-07-21
[Bug]: Worker deadlocks when agent backgrounds a server via shell `&` (orphaned subshell holds stdout pipe open)


maximilianopodvisory:
# Bug Description

The agent's worker process enters a permanent Python-level deadlock when the LLM issues a terminal command that backgrounds a long-running process with shell `&` (most commonly: `cd X && node server.js &` to start a dev server for verification). The worker becomes unrecoverable — the kanban dispatcher's `pid_alive` keeps extending the claim forever, no error is raised, no retry happens. The only recovery is `kill -9 <worker_pid>` from the operator.

This is the same pipe-holding mechanism as #60485 (open, cron scheduler) and #17327 (closed, poll loop) — but in a third code path (`tools/terminal_tool.py` + the main conversation loop in `agent/conversation_loop.py`) that the existing mitigations don't cover.

**Reproduces deterministically** when the LLM's tool-call pattern happens to include `&`-backgrounded servers during smoke tests. Observed 4 times in a single afternoon of E2E testing.

# Steps to Reproduce

A minimal repro requires a Hermes worker + an LLM that emits the specific command pattern. The trigger is any command of the form `... && <server> &` (or multi-line equivalents):

```bash
# From the agent (LLM-emitted tool call), in a hermes kanban worker session:
$ cd /app/express-api && node server.js &>/tmp/server.log &
```

What bash actually does with this:
- Bash parses `&` as lower-precedence than `&&`, so the whole compound `(cd /app/express-api && node server.js &>/tmp/server.log)` is backgrounded as a subshell.
- The subshell runs `cd`, then runs `node server.js` in the foreground of the subshell.
- `node server.js` is a server — it never exits.
- The subshell never exits (it's `wait4`-ing on node).
- The subshell inherited the Hermes terminal tool's stdout pipe, so the pipe's write end stays open.

For empirical repro in this Docker project (full evidence at https://github.com/maximilianoleiva/coding-scrum-lan/tree/main/benchmarks/cache-reclaim-fix-validation-20260721-1537 — notional link; repo may be private, evidence included inline below):

1. `docker exec` into a running Hermes container.
2. Have the agent (via `hermes -p coder chat -q work kanban task ...`) implement a Node/Express API.
3. The agent writes the server code, then attempts to smoke-test it. ~50% of the time it issues `cd /app && node server.js &...&` to start the server in the background.
4. Worker deadlocks within 1-2 minutes of that tool call. Permanently.

[...]

PR #42278, opened 2026-06-08
fix(tools): restore separator after backgrounded compound rewrite

entropy-0x:
`_rewrite_compound_background` rewrites `A && B &` into `A && { B & }` to
avoid the subshell-wait process leak. When another statement follows the
backgrounded compound on the same line (`A && B & C`), the trailing `&`
was the only separator between the compound and `C`. The rewrite consumed
that `&` into the brace group and produced `A && { B & } C`. A brace group
must be terminated by `;`, `&`, `|`, a newline, or `)`/`}` before the next
command, so the result is a bash syntax error and the entire command fails
to run — neither `A`, `B`, nor `C` execute. The rewrite is applied by
default to every foreground command, so a valid command is silently
mangled into one that errors out.

This restores a `;` separator after the closing `}` when the suffix
resumes with command text. Only spaces and tabs are stripped before the
check; a newline already terminates the group, and an existing separator
(`;`, `&`, `|`, `)`, `}`) is left untouched, so previously-correct rewrites
are unchanged.

## What does this PR do?

Fixes a rewrite in `_rewrite_compound_background` that turned a valid
foreground command of the form `A && B & C` into the bash syntax error
`A && { B & } C`, causing the whole command to fail. A `;` is now inserted
after the brace group whenever a further statement follows on the same
line, while leaving commands that already end in a separator/newline
untouched.

## Related Issue

N/A

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `tools/terminal_tool.py`: in `_rewrite_compound_background`, insert a `;`
  after the rewritten `{ ... & }` brace group when the trailing suffix
  begins with command text rather than a separator/terminator.
- `tests/tools/test_terminal_compound_background.py`: add
  `TestTrailingStatementSeparator` for the string-level rewrites and
  `TestRewriteIsValidBash`, which runs the rewriter output through
  `bash -n` for parse validity plus one end-to-end execution check.

## How to Test

1. Run `pytest tests/tools/test_terminal_compound_background.py -q`.
2. Before the fix, `_rewrite_compound_background("echo hi && sleep 5 & echo done")`
[...]

PR #98230, opened 2026-08-30
fix(terminal): keep a statement after background & valid after brace-group rewrite (#98222)

### LINKED ISSUE #98222: [Bug]: execute_code remote kernel spawn always fails on Docker/SSH/Modal — _rewrite_compound_background corrupts commands with a statement after &

lEWFkRAD:
## What does this PR do?

`_rewrite_compound_background` wraps `A && B &` tails as `A && { B & }` to avoid the subshell-wait leak (#68915). When a statement *follows* the `&` (`A && B & C`), the source `&` was also the terminator that following statement relied on — after the rewrite the bare word is stranded after `}`:

```
A && B & echo x  ->  A && { B & } echo x   (bash: syntax error near unexpected token `echo')
```

This kills every Docker/SSH/Modal `execute_code` remote-kernel spawn: the template in `tools/code_kernel_remote.py` ends `... & echo PID:$!`, the syntax error swallows the PID, the spawn returns `None`, and every call falls open to the per-call path (no persistence, interpreter restart per call). Blast radius is any `A && B & C` command on those backends, since `Environment.execute()` applies the rewrite by default.

The fix emits `A && { B & } ; C` when a token follows the `&` — `;` closes the brace group's command list. When the `&` ended the list (newline/EOF follows), output is byte-identical to today, so existing behavior is untouched.

Fixing the rewriter (rather than opting the spawn site out of the rewrite) repairs the whole class for every caller, and keeps the subshell-wait protection on commands that have a trailing statement — the skip-rewrite alternative would silently re-open the leak for `A && <long-running> & echo x`.

## Related Issue

Fixes #98222

## Type of Change

## Changes Made

- `tools/terminal_tool.py` — `_rewrite_compound_background`: insert ` ;` between the closing brace group and the suffix when the suffix starts with a token before any newline. +11/−1.
- `tests/tools/test_terminal_compound_background_tail.py` — regression tests: the issue's kernel-spawn template stays valid; separator forms for `&&`/`||`/multi-statement tails; runtime check that `echo PID:$!` receives a numeric PID; end-of-list outputs unchanged (byte-exact); idempotence on the new output form; runtime semantics (`&&` short-circuit with trailing statement, background job does not block the shell).

## How to Test

1. Reproduce on main (unit-level, any platform with bash):
   ```python
   from tools.terminal_tool import _rewrite_compound_background as rw
   out = rw("cd /tmp && nohup server.py > log 2>&1 & echo PID:$!")
   # main: 'cd /tmp && { nohup server.py > log 2>&1 & } echo PID:$!'
   ```
   `bash -n -c "$out"` → `syntax error near unexpected token 'echo'` (the input itself parses fine).
2. On this branch, `bash -n` passes on every rewritten output, and the runtime behaves: `bash -c "$(python -c ...)"` prints `PID:<digits>` and `/proc/<pid>/cmdline` confirms the PID is the backgrounded job.
3. `pytest tests/tools/test_terminal_compound_background.py tests/tools/test_terminal_compound_background_tail.py -q` → 26 passed. On main, the 7 new `#98222` tests fail (red); on this branch all green.
4. Semantics verified by execution (Git bash 5.2.37, Windows 11): trailing `echo` sees `$!` of the job; `false && B & echo TAIL` → TAIL prints and B never starts; long-running `B` returns the shell in ~0.1s (subshell-wait class stays fixed); failing trailing statement propagates its own exit code.

### Code

- [x] I've run `pytest tests/ -q` and all tests pass (targeted: both compound-background files, `test_local_background_child_hang.py`, `test_process_registry.py` — the 9 `test_process_registry.py` failures are identical on unpatched main tip in the same venv, i.e. pre-existing on Windows, unrelated to this change)
- [x] I've tested on my platform: Windows 11 (Git bash 5.2.37; the bug's backends run Linux bash — the `; ` separator is the canonical POSIX list terminator, verified with `bash -n` grammar and runtime behavior)

- [x] I've updated relevant documentation (README, `docs`, docstrings) — or N/A
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — the rewriter runs on all backends; output change applies uniformly and is bash-grammar-neutral (`;` terminates a list on Linux/macOS bash too)

### DIFF
--- a/tests/tools/test_terminal_compound_background_tail.py
+++ b/tests/tools/test_terminal_compound_background_tail.py
@@ -1,0 +1,108 @@
+"""Regression tests for #98222: a statement following the background `&`
+(`A && B & C`) must survive the brace-group rewrite as valid bash.
+
+Context: the rewriter turned `A && B &` into `A && { B & }` — correct when
+the `&` ends the list, but the source `&` was also the *terminator* for
+whatever followed it. `A && { B & } echo x` is a syntax error, so the
+remote-kernel spawn template (`... & echo PID:$!`) died on every Docker /
+SSH / Modal `execute_code` call. The fix emits `A && { B & } ; echo x`.
+"""
+
+import shutil
+import subprocess
+
+import pytest
+
+from tools.terminal_tool import _rewrite_compound_background as rewrite
+
+_BASH = shutil.which("bash")
+
+
+def _valid_bash(cmd: str) -> bool:
+    return subprocess.run(
+        ["bash", "-n", "-c", cmd], capture_output=True, text=True
+    ).returncode == 0
+
+
+needs_bash = pytest.mark.skipif(_BASH is None, reason="bash not available")
+
+
+class TestStatementAfterBackground:
+    """`A && B & C` — the `&` terminated BOTH the job and the list."""
+
+    def test_kernel_spawn_template_stays_valid(self):
+        cmd = (
+            "cd /tmp/x && nohup env A=B python3 kernel_runner.py "
+            "> /tmp/x/runner.log 2>&1 & echo PID:$!"
+        )
+        out = rewrite(cmd)
+        assert "PID:$!" in out
+        if _BASH:
+            assert _valid_bash(out), out
+
+    def test_echo_after_amp_gets_separator(self):
+        assert rewrite("A && B & echo done") == "A && { B & } ; echo done"
+
+    def test_or_chain(self):
+        assert rewrite("A || B & echo done") == "A || { B & } ; echo done"
+
+    def test_statement_with_chain_after_amp(self):
+        assert rewrite("A && B & C && D") == "A && { B & } ; C && D"
+
+    def test_pid_output_after_fix_is_numeric(self):
+        # Real runtime: the echo must print the backgrounded job's PID.
+        out = rewrite(
+            "true && sleep 45 >/dev/null 2>&1 & echo GOTPID:$!"
+        )
+        r = subprocess.run(
+            ["bash", "-c", out], capture_output=True, text=True, timeout=15
+        )
+        subprocess.run(
+            ["bash", "-c", "pkill -f 'sleep 45' 2>/dev/null; true"], timeout=10
+        )
+        token = next((t for t in r.stdout.split() if t.startswith("GOTPID:")), "")
+        assert token[7:].isdigit(), f"rc={r.returncode} out={r.stdout!r} err={r.stderr!r}"
+
+
+class TestEndOfListUnchanged:
+    """When nothing follows the `&`, the old output form must not change."""
+
+    def test_end_of_string(self):
+        assert rewrite("A && B &") == "A && { B & }"
+
+    def test_newline_after_amp(self):
+        assert rewrite("A && B &\nnext") == "A && { B & }\nnext"
+
+    def test_multiline_mixed(self):
+        cmd = "A && B & echo x\nC && D &"
+        assert rewrite(cmd) == "A && { B & } ; echo x\nC && { D & }"
+
+
+@needs_bash
+class TestRuntimeSemantics:
+    def test_background_does_not_block_shell(self):
+        # The original subshell-wait bug class must stay fixed.
+        out = rewrite("true && sleep 30 >/dev/null 2>&1 &")
+        subprocess.run(["bash", "-c", out], capture_output=True, timeout=8)
+        subprocess.run(
+            ["bash", "-c", "pkill -f 'sleep 30' 2>/dev/null; true"], timeout=10
+        )
+
+    def test_and_semantics_preserved_with_trailing_stmt(self):
+        # A fails -> B never starts, but the trailing echo still runs.
+        out = rewrite("false && sleep 97 >/dev/null 2>&1 & echo SKIPPED")
+        r = subprocess.run(
+            ["bash", "-c", out], capture_output=True, text=True, timeout=10
+        )
+        chk = subprocess.run(
+            ["bash", "-c", "pgrep -f 'sleep 97' || true"],
+            capture_output=True, text=True, timeout=10,
+        )
+        assert r.stdout.strip() == "SKIPPED"
+        assert chk.stdout.strip() == ""
+
+
+class TestIdempotence:
+    def test_rewriter_output_is_fixed_point(self):
+        once = rewrite("A && B & echo x")
+        assert rewrite(once) == once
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #41368: this uses the same semicolon-restoration mechanism in _rewrite_compound_background for same-line background tails. #41368 is the earliest open canonical PR.

kshitijk4poor:
Fixed via #101662 (salvage of #42278, which handles the `;;` / `;` / `|` suffix cases that a blanket `;` insertion breaks). Same bug, your PR confirmed it independently — credited in the salvage body. Closing as superseded; thanks.

PR #98231, opened 2026-08-30
fix(gateway): resend complete final when a stuck cursor freezes the partial

zhongwang:
## Problem

On Slack (edit-based streaming), a failed final edit can leave the streaming
cursor **stuck** on the visible partial: the user sees `"...scoring pass ▉"`
(truncated — the final `.` is missing), and the fallback then posts a **lone
`.`** as a fresh message while the frozen partial stays on screen. The result
is the "duplicated messages" symptom where the first message is missing its
last word.

## Root cause

`_continuation_text()` computes the tail from `_visible_prefix()`. When the
last edit failed, `_last_sent_text` is `"...scoring pass ▉"` — `_visible_prefix()`
strips the glyph but **keeps the cursor's leading space**, so the prefix is
`"...scoring pass "` and `final_text` (`"...scoring pass."`) starts with it →
continuation is the tail `"."`, posted as a new message. The frozen partial
(`...scoring pass` + stuck `▉`) is never cleaned up (the delete gate only
fires on a full-text resend).

## Fix

`_continuation_text()` now treats a visible prefix that ends with the cursor
as **untrustworthy**: the cursor is a synthetic tail that is not real
content. When `_last_sent_text` ends with the cursor (and the cursor-stripped
text is not the complete final), return the **full** `final_text` so the
fallback re-sends the complete answer and the existing full-resend delete
path removes the stale partial.

- Normal (cursor-free) partials: unchanged tail-only behavior.
- Stuck-cursor partial: full resend + delete → user sees one complete message.
- Split-turn reconcile / stale-finalize paths untouched (tests green).

## Tests

Added `test_fallback_final_full_resend_on_stuck_cursor_partial`, reproducing
the exact `"...scoring pass ▉"` → `"...scoring pass."` scenario and asserting
the complete final is re-sent and the partial deleted.

`tests/gateway/test_stream_consumer.py` (incl. new test) and
`tests/gateway/test_stale_finalize_suppression.py`: **83 passed**. Full
`tests/gateway/`: 6614 passed; 7 failures are pre-existing order-dependent
flakes (Discord send / session-store / Teams dotenv / Telegram polling),
unrelated to this change.

### DIFF
--- a/gateway/stream_consumer.py
+++ b/gateway/stream_consumer.py
@@ -2088,8 +2088,33 @@ class GatewayStreamConsumer:
         return self._clean_for_display(prefix)
 
     def _continuation_text(self, final_text: str) -> str:
-        """Return only the part of final_text the user has not already seen."""
+        """Return only the part of final_text the user has not already seen.
+
+        A visible prefix that ends with the streaming cursor is not reliable
+        delivered content: the cursor glyph is a synthetic tail that is not
+        part of ``final_text``, and a stuck cursor (left by a failed final
+        edit) means the user saw ``...pass`` while ``final_text`` is
+        ``...pass.``.  Computing a tail from that prefix would post a lone
+        ``.`` as a fresh message while the frozen ``...pass`` (with its
+        visible ``▉``) stays on screen — the duplicate/truncated-symptom.
+        When the prefix ends with the cursor, return the full ``final_text``
+        so the caller's fallback re-sends the complete answer and cleans up
+        the stale partial (its full-resend delete path).
+        """
         prefix = self._fallback_prefix or self._visible_prefix()
+        # The stuck cursor is only visible in the RAW last-sent text enough
+        # to matter; _visible_prefix() already stripped the glyph, so check
+        # the raw text directly.
+        if (
+            self.cfg.cursor
+            and self._last_sent_text
+            and self._last_sent_text.endswith(self.cfg.cursor)
+            and final_text != self._clean_for_display(
+                self._last_sent_text[: -len(self.cfg.cursor)]
+            ).strip()
+        ):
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Targeted fix for Slack edit-streaming stuck-cursor duplicate symptom. Detects frozen partial (raw `_last_sent_text` ends with cursor glyph) and falls back to full-final resend so stale partial is deleted rather than leaving truncated head + lone tail.

**What it does**
- `gateway/stream_consumer.py:8` extends `_continuation_text` docstring and adds guard before `prefix`/`final_text.startswith` tail slicing: if `cfg.cursor && _last_sent_text.endswith(cursor) && final_text != _clean_for_display(_last_sent_text[:-len(cursor)]).strip()` → cursor is not real content (`"Live ... scoring pass ▉"` vs `"Live ... scoring pass."`) → return full `final_text`. Caller `_send_fallback_final` then posts complete answer and deletes `msg_partial` via its full-resend path.
- Test `tests/gateway/test_stream_consumer.py:49` seeds `consumer._last_sent_text="Live ... scoring pass \u2589"` + `_message_id`, asserts fallback posts `"Live ... scoring pass."` and never a lone `"."`, and `delete_message` once, `final_response_sent`.

**Non-blocking notes**
- Guard inspects `_last_sent_text` (raw) not `visible_prefix` because visible prefix already strips glyph — correct; strip comparison uses `.strip()` to tolerate trailing space before cursor.
- Fallback to full resend is safe even on false positive (e.g., genuine `final_text` ends with `▉`? unlikely) — at worst duplicates a bit more content but still deletes partial; no data loss.
- Tail-only path (`final_text.startswith(prefix) → lstrip`) unchanged for healthy case; existing `test_fallback_final_keeps_partial_after_tail_only_send` stays green.

PR #98232, opened 2026-08-30
fix(tools): actionable anti-retry guidance in read_terminal empty-buffer error

c-pompa:
## Problem

When no desktop terminal tab is open (or the read times out), `read_terminal` returns a bare `No in-app terminal is open, or the read timed out.` A model that doesn't know the tool's precondition reads this as transient and retries — sometimes in a loop — expecting different output. Retrying can never help: the tool only works while a terminal tab is open in the Hermes desktop app.

## Fix

Extend the empty-buffer error message to state the precondition, say plainly not to retry, and point at the right alternatives (`terminal` / `execute_code`) when the model needs to run or check commands itself. Message-only change; no behavior difference when a terminal is open.

## Testing

- `tests/tools/test_read_terminal_tool.py` — covers the empty-buffer message content and the JSON passthrough path (2 passed)

### DIFF
--- a/tests/tools/test_read_terminal_tool.py
+++ b/tests/tools/test_read_terminal_tool.py
@@ -1,0 +1,33 @@
+#!/usr/bin/env python3
+"""Tests for tools.read_terminal_tool error guidance."""
+
+import json
+import unittest
+
+from tools.read_terminal_tool import read_terminal_tool
+
+
+class TestReadTerminalEmptyBufferGuidance(unittest.TestCase):
+    def test_empty_read_error_is_actionable(self):
+        """An empty in-app buffer must tell the agent not to retry and what
+        to use instead — weak models otherwise poll read_terminal in a loop
+        (observed 14x in session 20260731_230301_2c274e)."""
+        result = read_terminal_tool(callback=lambda **_: "")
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Low-risk UX fix for `read_terminal` empty-buffer retry loops. Error now states precondition and instructs not to retry, steering weak models toward `execute_code`/`terminal` tools.

**What it does**
- `tools/read_terminal_tool.py:49` changes bare `"No in-app terminal is open, or the read timed out."` to `"... read_terminal only works while a terminal tab is open in the Hermes desktop app — do not retry it expecting a different result. If you need to run or check commands yourself, use the terminal or execute_code tools instead."`
- Adds `tests/tools/test_read_terminal_tool.py` asserting error contains `"No in-app terminal is open"`, `"do not retry"`, `"execute_code"` and that `callback=None` desktop-only error is unchanged.

**Non-blocking notes**
- Message length increase is acceptable — tool errors are shown to the model, not truncated UI; the explicit "do not retry" phrasing is model-steerable and matches observed 14x poll loop.
- No behavior change for successful reads or non-empty payloads.

No blocking issues.

PR #98233, opened 2026-08-30
fix(telegram): make edit_message/delete_message replacement-aware

### LINKED ISSUE #98228: [Bug]: Telegram in-flight progress and cleanup use retired adapter after replacement reconnect
## Bug Description

When the Telegram reconnect watcher replaces a disconnected adapter during an active turn, the turn can retain the retired adapter for progress edits and post-delivery cleanup.

`send()` already delegates to a connected replacement adapter after #94498. `edit_message()` and `delete_message()` do not. On the retired adapter:

- `edit_message()` returns `SendResult(success=False, error="Not connected", retryable=False)`. The progress sender treats that as a permanent edit failure and sends later progress and heartbeat updates as new messages.
- `delete_message()` returns `False`. Post-delivery cleanup therefore leaves the temporary progress and heartbeat messages behind.

A long-running turn can consequently fragment into many Telegram messages after a successful reconnect even though progress grouping is configured to accumulate and cleanup is enabled.

## Steps to Reproduce

On current upstream `main` at `1a47a36422ae3c42c23839b72cc24fb6a651aa28`:

```python
import asyncio
from types import SimpleNamespace
from unittest.mock import AsyncMock

from gateway.config import PlatformConfig
from gateway.platforms.base import SendResult
from plugins.platforms.telegram.adapter import TelegramAdapter


async def main():
    old = TelegramAdapter(PlatformConfig(enabled=True, token="test-token"))
    live = TelegramAdapter(PlatformConfig(enabled=True, token="test-token"))
    old._bot = None
    live._bot = object()
    live.edit_message = AsyncMock(
        return_value=SendResult(success=True, message_id="7")
    )
    live.delete_message = AsyncMock(return_value=True)
    old.gateway_runner = SimpleNamespace(adapters={old.platform: live})

    edit = await old.edit_message("123", "7", "updated")
    deleted = await old.delete_message("123", "7")
    print(edit.success, edit.error, edit.retryable)
    print(live.edit_message.await_count)
    print(deleted, live.delete_message.await_count)


asyncio.run(main())
```

Observed output:

```text
False Not connected False
0
False 0
```

A production occurrence followed this sequence:

1. A long Telegram-triggered run was actively editing accumulated progress.
2. Telegram transport recovery rebuilt the adapter after `updater.stop()` exceeded the network-recovery deadline.
3. The background reconnect watcher installed a connected replacement adapter.
4. The active turn retained the retired adapter.
5. Ordinary sends delegated successfully, but progress edits returned non-retryable `Not connected` and cleanup deletion returned `False`.
6. Later progress and heartbeat updates became separate messages and remained after the final response.

## Expected Behavior

An in-flight turn should remain transport-correct after adapter replacement:

- Progress and heartbeat edits should resolve or delegate to the connected replacement adapter.
- Successful final delivery should delete all tracked temporary progress messages through the connected replacement.
- The final response should be delivered exactly once.
- Failed runs should continue retaining progress breadcrumbs.
- Permanent authentication failures should fail closed rather than waiting or delegating across an incompatible identity.

## Actual Behavior

`TelegramAdapter._replacement_telegram_adapter()` is used by `send()` only. The active turn's retired adapter still handles `edit_message()` and `delete_message()` directly. Since its `_bot` is `None`, editing fails as permanently non-retryable and deletion silently fails.

## Affected Component

Gateway (Telegram)

## Messaging Platform

Telegram

## Debug Report

A public debug bundle is intentionally omitted because it would include unrelated private configuration and logs. The deterministic reproduction above uses no credentials, customer data, or private topology. Sanitized production evidence showed the replacement reconnect succeeding before the stale edit and cleanup operations failed.

## Environment

- Operating system: macOS 26.5.2 arm64
- Hermes runtime: v0.20.6 (2026.8.27)
- Python runtime: 3.11.15
- Reproduced against upstream `main`: `1a47a36422ae3c42c23839b72cc24fb6a651aa28`

## Root Cause Analysis

The reconnect watcher replaces the adapter object in `runner.adapters`, but active turn closures retain the old object. Replacement awareness is implemented per adapter method rather than as a lifecycle-wide invariant:

- #91834 made replacement after a cancellation-shielded Telegram stop a supported recovery path.
- #94498 made ordinary `send()` replacement-aware.
- `edit_message()` and `delete_message()` remained replacement-unaware.

This is distinct from #27828 / #28485. That work preserves editing after an ordinary retryable network failure on the same adapter. Here the active object has been retired and returns a non-retryable `Not connected` before making any Telegram request.

## Related Work

- #95624 extends reconnect waiting/delegation to `edit_message()` and other outbound methods. It is the preferred implementation path for the edit side of this defect, but its current diff does not cover `delete_message()` or a full progress-to-cleanup adapter-swap regression.
- #27828 / #28485 cover retryable transient edit failures on the same adapter, not stale references after object replacement.
- #91834 established fresh-adapter replacement when Telegram shutdown hangs.
- #94498 added replacement delegation for `send()`.
- #47858 tracks broader delayed cleanup behavior. This issue concerns cleanup correctness during an already-supported replacement reconnect.

## Proposed Fix

Make the Telegram adapter contract replacement-aware for the complete in-flight lifecycle rather than adding separate gateway lookups at every progress caller:

1. Reuse the reconnect/delegation guard from #95624 for `edit_message()`.
2. Make `delete_message()` delegate to the connected replacement while preserving its `bool` contract.
3. Require the replacement to match the same Telegram platform/profile/credential context.
4. Add an integration regression that swaps the adapter during one active turn and verifies:
   - progress remains accumulated,
   - heartbeat edits use the live adapter,
   - successful cleanup deletes every tracked temporary message,
   - failed-run breadcrumbs remain,
   - permanent auth failures fail closed,
   - final delivery occurs exactly once.

bsdnn:
## What does this PR do?

When the reconnect watcher retires a disconnected Telegram adapter and installs
a connected replacement in `runner.adapters`, `send()` already delegates to it
(#94498) — but `edit_message()` and `delete_message()` still ran on the retired
adapter (`_bot is None`): edits returned a non-retryable "Not connected" and
deletes returned `False`. In-flight progress/heartbeat edits then fragmented
into fresh messages, and post-delivery cleanup left temporary messages behind
after a successful reconnect (#98228).

Fix: route both through the same replacement-awareness as `send()` — delegate to
the live adapter when present, otherwise wait for reconnection, and fail closed
on a permanent (auth) fatal instead of delegating across an incompatible
identity.

Credit to @0xble for the deterministic repro and diagnosis in #98228.

## Related Issue

Fixes #98228

## Type of Change

## Changes Made

- `plugins/platforms/telegram/adapter.py` — `edit_message`/`delete_message`
  delegate to the replacement adapter, mirroring `send()`.
- `tests/gateway/test_telegram_send_reconnect_wait.py` — edit/delete delegation
  + permanent-fatal fail-closed tests.

## How to Test

    pytest tests/gateway/test_telegram_send_reconnect_wait.py -q

8 pass. The issue's standalone repro now delegates to the live adapter
(edit → success, delete → True) instead of failing "Not connected".

## Checklist

### DIFF
--- a/plugins/platforms/telegram/adapter.py
+++ b/plugins/platforms/telegram/adapter.py
@@ -5670,7 +5670,30 @@ class TelegramAdapter(BasePlatformAdapter):
         edits target the most recent visible message.
         """
         if not self._bot:
-            return SendResult(success=False, error="Not connected")
+            # The reconnect watcher may have retired this adapter and installed
+            # a connected replacement in runner.adapters. Mirror send()'s
+            # replacement-awareness (#94498) so in-flight progress/heartbeat
+            # edits resolve through the live adapter instead of failing as a
+            # permanent "Not connected" and fragmenting into fresh messages
+            # (#98228).
+            live = self._replacement_telegram_adapter()
+            if live is not None:
+                return await live.edit_message(
+                    chat_id, message_id, content, finalize=finalize, metadata=metadata,
+                )
+            if self._is_permanent_fatal() or not await self._wait_for_reconnection():
+                return SendResult(
+                    success=False,
+                    error="Not connected",
+                    retryable=not self._is_permanent_fatal(),
+                )
+            live = self._replacement_telegram_adapter()
+            if not self._bot and live is not None:
+                return await live.edit_message(
+                    chat_id, message_id, content, finalize=finalize, metadata=metadata,
+                )
+            if not self._bot:
+                return SendResult(success=False, error="Not connected", retryable=True)
 
         # Rich finalize (Bot API 10.1): when the completed content has
         # constructs the legacy MarkdownV2 edit degrades (tables → bullet
@@ -6081,7 +6104,20 @@ class TelegramAdapter(BasePlatformAdapter):
         caller leaves the preview in place and logs at debug level.
         """
         if not self._bot:
-            return False
+            # Retired adapter after a replacement reconnect: delegate cleanup to
+            # the live adapter so post-delivery deletion of temporary progress/
+            # heartbeat messages still succeeds (#98228). Mirrors send()/
+            # edit_message() replacement-awareness.
+            live = self._replacement_telegram_adapter()
+            if live is not None:
+                return await live.delete_message(chat_id, message_id)
+            if self._is_permanent_fatal() or not await self._wait_for_reconnection():
+                return False
+            live = self._replacement_telegram_adapter()
+            if not self._bot and live is not None:
+                return await live.delete_message(chat_id, message_id)
+            if not self._bot:
+                return False
         try:
             await self._bot.delete_message(
                 chat_id=normalize_telegram_chat_id(chat_id),
--- a/tests/gateway/test_telegram_send_reconnect_wait.py
+++ b/tests/gateway/test_telegram_send_reconnect_wait.py
@@ -14,6 +14,7 @@ from unittest.mock import AsyncMock, MagicMock
 import pytest
 
 from gateway.config import PlatformConfig
+from gateway.platforms.base import SendResult
 from plugins.platforms.telegram.adapter import TelegramAdapter  # noqa: E402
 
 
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Mirrors `send()` replacement-awareness (#94498) onto `edit_message`/`delete_message`, preventing "Not connected" fragmentation after reconnect watcher swaps adapter.

**What it does**
- `plugins/platforms/telegram/adapter.py:5670` `edit_message` and `6081` `delete_message`: when `not self._bot`, first checks `_replacement_telegram_adapter()` → delegate immediately to live adapter; otherwise if `permanent_fatal` or `await _wait_for_reconnection()` fails → fail with `retryable=False/True` accordingly; re-checks live after wait and delegates, else final `Not connected`.
- Tests `tests/gateway/test_telegram_send_reconnect_wait.py:76` assert `edit_message`/`delete_message` delegate to `live` when already live (no wait), and permanent fatal fails closed without wait/delegation.

**Non-blocking notes**
- Delegation preserves `finalize`/`metadata`/`chat_id` args verbatim — correct for in-flight progress/heartbeat edits that must land on the latest `message_id`.
- `_wait_for_reconnection` path is bounded by reconnect watcher (not infinite); `retryable` flag now correctly set on final `Not connected` (previously implicit) — callers can distinguish transient vs auth fatal.
- Ensure `_replacement_telegram_adapter()` identity check avoids delegation across incompatible `platform` keys (it checks `runner.adapters[platform]`).

PR #98235, opened 2026-08-30
fix(telegram): use cached config during adapter rebuild

yazan1991:
## Summary

Use the gateway's existing cached config loader when resolving Telegram notification mode during adapter construction, instead of reparsing the full gateway YAML configuration synchronously.

## Problem

Telegram adapter rebuilds can happen from the gateway reconnect path. `_resolve_notifications_mode()` currently calls `gateway.config.load_gateway_config()`, which opens and parses the full gateway config with PyYAML synchronously on the gateway event-loop thread.

This behavior was introduced as part of the platform plugin migration in `5600105478ffde29d7566b45421b100eaa29c4ef`. Before that migration, this path used the gateway's cached config semantics.

Under reconnect/rebuild conditions, a synchronous full YAML parse can stall the event loop and delay gateway health handling and recovery work.

## Fix

Restore the cached loader semantics by using the existing `gateway.run._load_gateway_config()` helper from `_resolve_notifications_mode()`.

The runtime change is intentionally small: 2 additions and 3 deletions.

## Tests

Adds focused regression coverage verifying that:

- Telegram adapter construction uses the cached gateway config and does not invoke the full config loader.
- `HERMES_TELEGRAM_NOTIFICATIONS` retains precedence.
- Invalid cached notification modes continue to fail closed to `important`.
- A missing notification mode preserves the existing `important` default.

Focused test run: 100 passed, 2 skipped, 0 failed. Ruff, formatting, compile, and diff checks also pass.

### DIFF
--- a/plugins/platforms/telegram/adapter.py
+++ b/plugins/platforms/telegram/adapter.py
@@ -10841,9 +10841,8 @@ def _resolve_notifications_mode() -> str:
     mode = os.getenv("HERMES_TELEGRAM_NOTIFICATIONS", "")
     if not mode:
         try:
-            from gateway.config import load_gateway_config
-            from gateway.run import cfg_get
-            _gw_cfg = load_gateway_config()
+            from gateway.run import _load_gateway_config, cfg_get
+            _gw_cfg = _load_gateway_config()
             _raw = cfg_get(_gw_cfg, "display", "platforms", "telegram", "notifications")
             if _raw not in {None, ""}:
                 mode = str(_raw).strip().lower()
--- a/tests/gateway/test_telegram_adapter_build_config.py
+++ b/tests/gateway/test_telegram_adapter_build_config.py
@@ -1,0 +1,60 @@
+from gateway.config import PlatformConfig
+from plugins.platforms.telegram.adapter import (
+    _build_adapter,
+    _resolve_notifications_mode,
+)
+
+
+def test_adapter_build_uses_cached_gateway_config_not_full_loader(monkeypatch):
+    import gateway.config as gateway_config
+    import gateway.run as gateway_run
+
+    config = PlatformConfig(enabled=True, token="test-token")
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Small latency/safety fix — Telegram adapter rebuild now reads notifications mode from the gateway's cached config loader rather than reparsing YAML synchronously on the event loop.

**What it does**
- `plugins/platforms/telegram/adapter.py:10841` `_resolve_notifications_mode()` changes `from gateway.config import load_gateway_config` → `from gateway.run import _load_gateway_config` + `cfg_get`; uses `_load_gateway_config()` (cached) instead of `load_gateway_config()` (full PyYAML parse).
- Adds `tests/gateway/test_telegram_adapter_build_config.py` asserting `_build_adapter` uses `_load_gateway_config` not `load_gateway_config` (patched to assert), env override precedence preserved, invalid/cached missing mode fails closed to `"important"` with log.

**Non-blocking notes**
- `_load_gateway_config` is intentionally private (`_`) — coupling to `gateway.run` is documented ("existing cached loader"); if `gateway.run` refactors, this import will break loudly — acceptable given rebuild happens on event-loop thread where blocking parse was observed (link in body).
- Behavior otherwise identical — no new config key, no fallback change.

No blocking issues.

kshitijk4poor:
Thanks @yazan1991 for the work here. I reviewed it against current `main` in a perf sweep and I'm closing it:

main plugins/platforms/telegram/adapter.py::_resolve_notifications_mode L7130-7137 still uses gateway.config.load_gateway_config(); but gateway.run._load_gateway_config L2880 is NOT a cache on main — it calls load_user_config_effective each call. _resolve_notifications_mode runs once per adapter build (connect/rebuild), not per message.

If you can show a measured cost on a hot path in a current build, a focused PR on top of today's `main` is welcome.

PR #98240, opened 2026-08-30
Reduce unchanged Kanban notifier polling I/O

royalaid:
## Summary

A 45-second Microsoft Defender Performance Analyzer capture recorded 1,127 scans of `kanban.db-shm` across 10 board paths, with 12.1 seconds of cumulative Defender scan time. The Kanban notifier was reopening and rescanning unchanged board databases on every five-second tick.

This change adds a watcher-lifetime cache keyed by the resolved board path and notifier context. Before querying, it compares documented filesystem metadata for the database, `-wal`, and `-shm` files (presence, device/inode, size, mtime, and ctime). A stable successful query is cached; unchanged subsequent ticks skip SQLite opens. The cache is invalidated by any tracked file change, profile/adapter context change, board replacement, or board lifecycle change.

## Correctness and trade-off

- The implementation does not depend on undocumented SQLite WAL internals; it uses the documented `os.stat` filesystem API and tracks all three SQLite files.
- The signature is checked before and after a query. If a writer races the query, the result is not cached and the next tick queries again.
- Any stat or query uncertainty fails open to the existing database query path rather than suppressing notifications.
- Subscription add/remove, WAL commits, checkpoint/truncation, replacement/recreation, board deletion, multiple boards/profiles, concurrent writers, and normal shutdown are covered by deterministic tests.
- As with any metadata-based optimization, a filesystem that reports no distinguishable metadata change for an in-place write can reduce invalidation fidelity. Tracking DB/WAL/SHM metadata, using nanosecond timestamps where available, checking before/after stability, and failing open on probe errors limit that risk. No new write, locking, or authorization bypass is introduced.

## Measured behavior

- Regression on the recorded upstream baseline: 3 unchanged poll ticks opened SQLite 6 times; the new counter test failed against the target expectation of 2 opens.
- Fix: the same scenario opens SQLite 2 times total (initial read-only subscription probe plus board load), then performs zero repeated opens on the two unchanged ticks.
- A committed board change is observed on the next poll within the existing interval; the fixture includes a WAL commit and `PRAGMA wal_checkpoint(TRUNCATE)` and delivers the event.

## Verification

Passed on the rebased branch:

```text
HERMES_PYTHON=<managed interpreter> scripts/run_tests.sh tests/gateway/test_kanban_notifier.py tests/gateway/test_kanban_notifier_io.py tests/gateway/test_kanban_notifier_apiserver_wake.py tests/gateway/test_kanban_notifier_wake_only_ordering.py tests/gateway/test_kanban_notifier_watcher_dispatch_gate.py tests/gateway/test_kanban_notifier_zero_sub_gate.py tests/gateway/test_kanban_changes_requested_notifier.py -q
36 passed, 0 failed

uvx --from ruff==0.15.10 ruff check gateway/kanban_watchers.py tests/gateway/test_kanban_notifier.py tests/gateway/test_kanban_notifier_io.py
All checks passed

python -m compileall -q gateway/kanban_watchers.py tests/gateway/test_kanban_notifier.py tests/gateway/test_kanban_notifier_io.py
exit 0
git diff upstream/main --check
exit 0
```

An attempted unbounded `tests/gateway -q` fanout exceeded its host timeout and is not claimed as a pass; the focused canonical runner above is the acceptance result.

## Rollback

Revert commit `3c2f17ceab8f067b9d9ccaf6cb3108909ea132b3` to restore the prior polling behavior.


kshitijk4poor:
From the perf sweep: this isn't in the kanban stack (#120408) this round. A rebase onto current main left one notifier test failing, and I ran out of time to settle it properly. It stays open for the next pass.

PR #98241, opened 2026-08-30
fix(slack): do not wake on unmentioned bot messages in force-mention channels

MaverickKB:
## Problem

A channel listed in `require_mention_channels` always requires an explicit bot @mention. However, an unmentioned message from a **peer bot** still auto-triggered a turn whenever a wake check approved it (previously mentioned thread, active session). In a coordination channel shared with other bots this made the adapter a free-for-all: every peer bot status post or acknowledgement inside an ongoing thread started a new agent run and burned model tokens.

## Change

In a force-mention channel, an unmentioned bot-authored message is dropped **before** the wake checks:

- `sender_is_bot_user` is now always bound (defaults to `False`) so the gate can rely on it.
- A new branch runs only when the channel is in `require_mention_channels` **and** the message is unmentioned **and** the sender is a bot: the message is ignored.
- Human senders keep the existing wake-check auto-follow contract (previously mentioned thread, active session).
- A direct @mention still reaches the bot regardless of sender.

## Why this scope

The fix is deliberately narrow. It does not change global behaviour, it does not affect channels without the `require_mention_channels` override, and it preserves the documented contract that wake checks still apply to human-initiated follow-ups. It only stops one class of unwanted turn: unmentioned, bot-authored traffic in channels that the operator has already declared "mention required."

## Tests

- New: `test_forced_channel_ignores_unmentioned_bot_message` — unmentioned peer bot message in a previously mentioned thread is ignored.
- Existing: `test_forced_channel_wake_checks_still_apply` — human thread follow-up still processed.
- Full Slack suite: 238 tests passed (test_slack.py + test_slack_require_mention_channels.py).
- Adjacent contract suites (ignore_other_user_mentions, wake_external_bot_messages, mention): 41 tests passed.
- `ruff check` clean; `py_compile` clean.

### DIFF
--- a/plugins/platforms/slack/adapter.py
+++ b/plugins/platforms/slack/adapter.py
@@ -6443,6 +6443,11 @@ class SlackAdapter(BasePlatformAdapter):
                     return
                 if allow_bots == "mentions" and not is_mentioned:
                     return
+        else:
+            # A message with no user id, or whose user is this bot, is not
+            # bot-to-bot traffic. Default the sender flag to False so the
+            # force-mention channel check below can rely on it being bound.
+            sender_is_bot_user = False
 
         if not is_one_to_one_dm and bot_uid:
             # Check allowed channels — if set, only respond in these channels (whitelist)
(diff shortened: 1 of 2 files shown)


srrevusfaggart-collab:
Stop it

srrevusfaggart-collab:
Some one getting fired

PR #95954, opened 2026-08-27
fix(gateway): block direct-spawn self-lifecycle

giofxjitas:
## What does this PR do?

Prevents a live Windows gateway started via direct spawn from stopping or restarting itself through the `terminal` or `execute_code` tools.

The lifecycle guards previously relied on `_is_supervised_gateway_process()`. A direct-spawn Windows gateway owns the live gateway PID file but has no systemd/launchd supervisor, so the guard returned false. If an agent then ran `hermes gateway restart` through its own terminal tool, the gateway stopped itself and interrupted the tool before the replacement process could be launched.

This separates **live gateway identity** from **external supervision**:

- `_is_gateway_process()` verifies `_HERMES_GATEWAY=1` plus ownership of the live gateway PID file.
- `_is_supervised_gateway_process()` builds on that identity and separately checks for a supervisor.
- The `terminal` and `execute_code` lifecycle guards use live gateway identity, covering direct-spawn Windows gateways without treating descendants that merely inherit `_HERMES_GATEWAY` as the gateway.

## Related Issue

No matching issue or PR was found in a repository search.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `tools/process_registry.py`: distinguish live gateway PID ownership from external supervision.
- `tools/terminal_tool.py`: block gateway lifecycle commands from any live gateway, including Windows direct-spawn processes.
- `tools/code_execution_tool.py`: apply the same guard to lifecycle commands embedded in generated code.
- `tests/hermes_cli/test_gateway_restart_loop.py`: add direct-spawn identity and terminal/execute-code regressions while preserving inherited-environment CLI behavior.

## How to Test

1. Run the focused regression set:
   ```bash
   uv run --isolated --with pytest python -m pytest \
     tests/hermes_cli/test_gateway_restart_loop.py::TestGatewayProcessIdentity \
     tests/hermes_cli/test_gateway_restart_loop.py::TestExecuteCodeGatewayLifecycleGuard \
     tests/hermes_cli/test_gateway_restart_loop.py::TestTerminalToolGatewayLifecycleGuard::test_detached_gateway_blocks_lifecycle_command \
[...]

PR #98244, opened 2026-08-30
fix(gateway): block self-restart on a direct-spawn Windows gateway with no supervisor

### LINKED ISSUE #98221: [Bug]: Direct-spawn Windows gateway can stop itself via terminal restart and never relaunch
I hit this twice on Hermes 0.20.6 on Windows.

I was talking to Hermes through Telegram and approved its request to restart the gateway. Hermes ran `hermes gateway restart` using its terminal tool. The gateway shut down, but it never came back.

The log showed:

```text
Tool terminal returned error: [Command interrupted]
exit_code: 130
Gateway shutting down
Gateway stopped
```

After that, `hermes gateway status` reported that no gateway process was running. I had to open Command Prompt and run `hermes gateway start` myself.

Steps to reproduce:

1. Start Hermes Gateway on Windows using the installed Startup-folder VBS launcher, without a separate service supervisor.
2. Talk to Hermes through Telegram and ask it to restart.
3. Approve the terminal command when Hermes tries to run `hermes gateway restart`.
4. The gateway shuts down, interrupts its own terminal command, and never reaches the start step.

It looks like the restart command is running as a child of the gateway it is trying to stop. When the gateway shuts down, that child receives the interruption before it can launch the replacement process.

Expected behavior: Hermes should either use the safe built-in `/restart` path, or refuse to run a self-restart through its own terminal and tell the user to restart it externally.

Environment:

- Hermes 0.20.6
- Windows build 10.0.26200.9168
- Gateway launched through the installed Startup-folder VBS
- Request came from Telegram

Related work:

- #95954 described this exact Windows direct-spawn problem, but it was closed without being merged.
- #43969 fixed a different problem in the built-in Telegram `/restart` path.

I removed all tokens, profile details, chat IDs, and private paths from this report.

chelsealong:
## What does this PR do?

Fixes #98221: on Windows, a gateway started by direct spawn (the installed
Startup-folder VBS launcher, no systemd/launchd/schtasks supervisor) can be
told by its own agent to run `hermes gateway restart` through the `terminal`
tool. The command is never blocked, so it stops the live gateway — and the
terminal tool's own child process is interrupted along with it before the
replacement gateway can start. The gateway never comes back; the user has to
start it manually.

## Root cause

`tools/terminal_tool.py` and `tools/code_execution_tool.py` both hard-block
gateway lifecycle commands (`hermes gateway restart|stop|uninstall`,
`systemctl`/`launchctl` equivalents) when running inside the gateway process,
via `tools/process_registry.py::_is_supervised_gateway_process()`. That check
requires **both** live gateway PID-file ownership **and** an external
supervisor marker (`INVOCATION_ID` for systemd, `XPC_SERVICE_NAME` for
launchd, etc.).

A direct-spawn Windows gateway owns the PID file but has no supervisor
marker, so the guard always evaluated to `False` for it — the block never
fired, and the destructive command ran unguarded.

## Changes Made

- `tools/process_registry.py`: split live gateway identity from external
  supervision. New `_is_gateway_process()` checks only `_HERMES_GATEWAY=1` +
  PID-file ownership. `_is_supervised_gateway_process()` is now built on top
  of it with the additional supervisor check, preserving its existing
  behavior for the systemd-scope cgroup isolation path (the one caller that
  actually needs supervision, unchanged).
- `tools/terminal_tool.py`: the lifecycle hard-block now gates on
  `_is_gateway_process()` instead of `_is_supervised_gateway_process()`.
- `tools/code_execution_tool.py`: same change, mirroring the terminal_tool
  guard as the file's existing comment already says it should.
- `tests/hermes_cli/test_gateway_restart_loop.py`: retargeted the existing
  terminal-tool guard tests' monkeypatch to `_is_gateway_process` (no
  behavior change to those tests — they already modeled "is this process the
  gateway" as a single boolean); added `TestTerminalToolBlocksUnsupervisedDirectSpawnGateway`,
  which exercises the real identity primitives (not a patched boolean) to
  prove the guard now fires for a live, unsupervised gateway, and that this
  identity still does **not** satisfy the stricter supervised check other
  callers rely on.

## Scope note

This is deliberately narrower than the earlier attempt at this issue (#95954,
closed unmerged). That PR's review correctly endorsed the same live-identity/
supervision split, but also asked for a tri-state identity result and for
`hermes_cli/gateway.py`'s own `stop`/`restart`/`uninstall` self-target guards
to move from PID-equality to process-ancestry, to close a *separate* bypass
where dynamically-constructed code in `execute_code` avoids the source-text
classifier entirely. That is a real, independent hardening problem, but it is
not what #98221 reports (a literal `hermes gateway restart` invoked through
the `terminal` tool), and `hermes_cli/gateway.py`'s own self-checks are not
the code path this specific report exercises — the terminal-tool-spawned CLI
process has a different PID than the gateway, so PID-equality there behaves
identically before and after this change. Fixing the deeper `execute_code`
bypass is left as a separate, follow-up piece of work.

## How to Test

Reproduction (matches the issue): with a gateway running via direct spawn on
Windows (no supervisor), have the agent run `hermes gateway restart` through
the `terminal` tool. Before this fix, the command executes and the gateway
never comes back. After this fix, the tool call is blocked with:

```
Blocked: command or referenced script cannot restart, stop, or uninstall the
gateway from inside the gateway process. ...
```

Automated regression:
```
uv run --isolated --with pytest python -m pytest \
  tests/hermes_cli/test_gateway_restart_loop.py -q
```

Proved the new regression test fails without the fix: stashed the three
source files, re-ran the suite, and confirmed
`TestTerminalToolBlocksUnsupervisedDirectSpawnGateway::test_blocks_restart_with_no_supervisor_present`
fails (`assert -1 == 1` — the command ran instead of being blocked), along
with every other test that patches the identity function under its old name.
Restored the fix and reran.

### Test output (with the fix)

```
$ uv run --isolated --with pytest python -m pytest tests/hermes_cli/test_gateway_restart_loop.py -q
299 passed in 3.69s

$ uv run --isolated --with pytest --with pytest-asyncio python -m pytest tests/hermes_cli/ -q -k gateway
810 passed, 18 skipped, 6940 deselected in 58.38s

$ uv run --isolated --with pytest python -m pytest tests/tools/test_process_registry.py tests/tools/test_code_execution.py tests/tools/test_code_execution_modes.py tests/tools/test_execute_code_approval_cluster.py tests/tools/test_approval.py -q
297 passed, 4 skipped, 7 subtests passed in 15.83s

$ uv run --isolated --with ruff ruff check tools/process_registry.py tools/terminal_tool.py tools/code_execution_tool.py tests/hermes_cli/test_gateway_restart_loop.py
All checks passed!
```

## Type of Change

### Code

- [x] My commit message follows Conventional Commits
- [x] I searched for existing PRs/issues and found no duplicate (see Scope
      note above re: the related, closed #95954)
- [x] I've run `pytest tests/hermes_cli/test_gateway_restart_loop.py -q` and
      the related tool/process-registry suites, all pass
- [x] I've added tests for the bug
- [ ] I've tested on Windows — not available in this environment; validated
      via the identity primitives directly (`gateway.status.get_running_pid`,
      the supervisor-marker env vars) rather than an end-to-end Windows run

- [x] Documentation update — N/A; internal lifecycle safety behavior
- [x] `cli-config.yaml.example` update — N/A; no config keys changed
- [x] `CONTRIBUTING.md` / `AGENTS.md` update — N/A; no workflow/architecture
      contract changed
- [x] Cross-platform impact considered — this is specifically a Windows
      direct-spawn fix; POSIX supervised/unsupervised behavior is unchanged
      because `_is_gateway_process()` is a strict superset of what
      `_is_supervised_gateway_process()` already matched
- [x] Tool descriptions/schemas — N/A; no tool schema changed, only internal
      guard logic

## AI assistance disclosure

This PR was prepared with AI assistance (an autonomous coding agent), with
the root-cause analysis, scope decision, and test verification described
above performed and checked as part of that process.

### DIFF
--- a/tests/hermes_cli/test_gateway_restart_loop.py
+++ b/tests/hermes_cli/test_gateway_restart_loop.py
@@ -518,7 +518,7 @@ class TestTerminalToolGatewayLifecycleGuard:
         monkeypatch.setattr(tt, "_task_env_overrides", {})
         monkeypatch.setattr(tt, "_get_env_config", self._minimal_config)
         monkeypatch.setattr(
-            process_registry, "_is_gateway_process",
+            process_registry, "_is_gateway_process_or_unknown",
             lambda: inside_gateway,
         )
 
@@ -901,8 +901,54 @@ class TestTerminalToolBlocksUnsupervisedDirectSpawnGateway:
         monkeypatch.delenv("HERMES_S6_SUPERVISED_CHILD", raising=False)
         monkeypatch.delenv("HERMES_GATEWAY_EXTERNAL_SUPERVISOR", raising=False)
         monkeypatch.setattr(
-            "gateway.status.get_running_pid",
-            lambda *, cleanup_stale=False: os.getpid(),
+            "gateway.status.get_running_pid_identity_strict",
+            lambda pid_path: (os.getpid(), 123.0),
+        )
+
+        result = json.loads(tt.terminal_tool(command="hermes gateway restart"))
+
+        assert result["exit_code"] == 1
+        assert "Blocked" in result["error"]
+
+    def test_blocks_restart_on_unreadable_identity(self, monkeypatch):
+        """P1 (review, 2026-08-30): an unreadable PID/lock record is not
+        proof this process is NOT the gateway -- it is unproven. The hard
+        block must fail closed (still refuse) rather than let a transient
+        identity-read error grant the destructive command safe passage.
+        """
+        import tools.terminal_tool as tt
+
+        self._patch_env(monkeypatch, self._make_fake_env())
+
+        monkeypatch.setenv("_HERMES_GATEWAY", "1")
+
+        def _raise(pid_path):
+            raise OSError("PID file unreadable")
+
+        monkeypatch.setattr(
+            "gateway.status.get_running_pid_identity_strict", _raise
+        )
+
+        result = json.loads(tt.terminal_tool(command="hermes gateway restart"))
+
+        assert result["exit_code"] == 1
+        assert "Blocked" in result["error"]
+
+    def test_blocks_restart_on_malformed_identity_metadata(self, monkeypatch):
+        """P1 (review, 2026-08-30): an active-but-malformed PID/lock record
+        (the exact case ``get_running_pid_identity_strict`` raises for) must
+        also fail closed, not be read as "not the gateway"."""
+        import tools.terminal_tool as tt
+
+        self._patch_env(monkeypatch, self._make_fake_env())
+
+        monkeypatch.setenv("_HERMES_GATEWAY", "1")
+
+        def _raise(pid_path):
+            raise RuntimeError("gateway PID or lock metadata is malformed")
+
+        monkeypatch.setattr(
+            "gateway.status.get_running_pid_identity_strict", _raise
         )
 
         result = json.loads(tt.terminal_tool(command="hermes gateway restart"))
@@ -1913,7 +1959,7 @@ class TestTerminalToolGatewayLifecycleGuardRemote:
         monkeypatch.setattr(tt, "_task_env_overrides", {})
         monkeypatch.setattr(tt, "_get_env_config", lambda: {"env_type": "local", "cwd": "/tmp", "timeout": 60, "lifetime_seconds": 3600})
         monkeypatch.setattr(
-            process_registry, "_is_gateway_process",
+            process_registry, "_is_gateway_process_or_unknown",
             lambda: inside_gateway,
         )
 
--- a/tests/hermes_cli/test_gateway_restart_loop.py
+++ b/tests/hermes_cli/test_gateway_restart_loop.py
@@ -518,7 +518,7 @@ class TestTerminalToolGatewayLifecycleGuard:
         monkeypatch.setattr(tt, "_task_env_overrides", {})
         monkeypatch.setattr(tt, "_get_env_config", self._minimal_config)
         monkeypatch.setattr(
-            process_registry, "_is_supervised_gateway_process",
+            process_registry, "_is_gateway_process",
             lambda: inside_gateway,
         )
 
@@ -636,8 +636,8 @@ class TestTerminalToolGatewayLifecycleGuard:
         self, monkeypatch
     ):
         """#92560: CLI/TUI agent sessions inherit _HERMES_GATEWAY=1 from the
-        gateway but are NOT the gateway supervisor.  The env gate must not
-        fire for them — only for the actual gateway process (PID-file owner).
+        gateway but are not the gateway itself.  The env gate must not fire
+        for them — only for the actual gateway process (PID-file owner).
         """
         import tools.terminal_tool as tt
 
@@ -652,8 +652,8 @@ class TestTerminalToolGatewayLifecycleGuard:
 
         # Simulate a CLI agent session: _HERMES_GATEWAY=1 is in the
         # environment (inherited from the gateway), but
-        # _is_supervised_gateway_process() returns False because the
-        # process does not own the gateway PID file.
+        # _is_gateway_process() returns False because the process does not
+        # own the gateway PID file.
         self._patch_env(monkeypatch, _FakeEnv(), inside_gateway=False)
         monkeypatch.setenv("_HERMES_GATEWAY", "1")
         monkeypatch.setattr(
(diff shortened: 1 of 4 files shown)


andrexibiza (commented):
Reviewed the exact current head `ed887f2ff2b3b34cbdea99f4c405f533dff0a995`, the four-file diff, the full lifecycle/identity call chain, #98221, the closed predecessor #95954, the merged ownership lineage in #93267, current-main drift, open `terminal_tool.py` collisions, and the hosted checks.

The core decomposition here is correct: **“am I the live gateway?” and “am I externally supervised?” are different facts.** Keeping systemd-scope isolation on the stricter supervision predicate while moving the terminal hard block to live gateway identity is the right direction for the Windows direct-spawn failure. The new regression also exercises the real PID-owner predicate instead of proving only a monkeypatched boolean. Exact-head CI is strong: CI [33285358240](https://github.com/NousResearch/hermes-agent/actions/runs/33285358240), Docker [33285357987](https://github.com/NousResearch/hermes-agent/actions/runs/33285357987), and Nix [33285357988](https://github.com/NousResearch/hermes-agent/actions/runs/33285357988) all succeeded; the CI run includes Python, e2e, Windows-only, macOS-only, blocking Ruff/footguns, contributor attribution, and `All required checks pass`.

I still see two landing blockers.

### P1 — the new destructive-operation identity predicate still fails open on an indeterminate PID/lock read

`tools/process_registry.py::_is_gateway_process()` now owns the decision that gates the terminal and `execute_code` lifecycle hard blocks. Its shape is:

```py
if os.environ.get("_HERMES_GATEWAY") != "1":
    return False
try:
    from gateway.status import get_running_pid
    return get_running_pid(cleanup_stale=False) == os.getpid()
except Exception:
    return False
```

That final `False` is not “proved not to be the gateway”; it is **“could not prove identity.”** Both changed consumers interpret it as permission to continue, so a transient/unreadable/malformed gateway identity record turns an uncertainty at the safety check into permission for the exact destructive command this PR is trying to refuse. This is inside #98221's literal `terminal -> hermes gateway restart` path, not the separately scoped dynamic-`execute_code` bypass.

There is already a stricter identity primitive in `gateway/status.py`: `get_running_pid_identity_strict()` distinguishes an inactive runtime from an active-but-ambiguous one and raises for states such as an active gateway lock with missing PID metadata or malformed PID/lock metadata. That is the semantic distinction the hard block needs. A destructive lifecycle gate should have at least `self / other-or-inactive / unknown`, with `unknown` refusing pre-effect; it must not collapse `unknown` into the allow case.

Required repair: make the live-gateway identity check preserve uncertainty, and make both changed hard-block consumers fail closed on that uncertainty. Add zero-effect regressions for at least (a) PID/lock identity read failure and (b) active-but-malformed identity metadata, proving no terminal environment execution / code-execution child is started. Keep the normal non-gateway allow case and the direct-spawn positive case so the fix does not regress ordinary CLI children.

This also matches the wider repo's own mutation-boundary behavior: `hermes_cli/gateway.py`'s liveness code already treats witness conflict/unknown as non-authority for destructive escalation rather than turning an unreadable proof into permission.

### P1 — #98244 is a superseding salvage of #95954, but the landing provenance does not preserve `giofxjitas`' authored mechanism

The relationship is stronger than “related earlier attempt.” Closed-unmerged #95954 by `giofxjitas` already introduced the same production mechanism on the same surfaces: `_is_gateway_process()` in `tools/process_registry.py`, terminal and `execute_code` switching from `_is_supervised_gateway_process()` to that predicate, and the supervision predicate becoming a refinement of live identity. #98244 then recreates that mechanism in a new `chelsealong` commit, with changed/expanded issue-focused regression coverage and comments.

That makes #98244 a **superseding salvage/convergence candidate**, not a duplicate and not provenance-free new work. The current body correctly points to #95954, but the commit history attributes the reimplemented mechanism only to the new submitter. Contributor-attribution CI being green does not resolve origin attribution; it checks mapped commit identities, not whether a superseded author's work disappeared from the landing lineage.

There is a good repository precedent immediately in this same defect family: merged #93267 explicitly salvaged #92196's guard work from `nbxuhk` and #92633's regression from `DavidMetcalfe`, preserving both authorships via cherry-picked commits. Please do the equivalent here: preserve `giofxjitas`' authored contribution in the landing object (preferably by retaining/cherry-picking the authored commit/range where composition permits, otherwise by an explicit commit-level co-author/origin record that survives squash/merge), while keeping `chelsealong`'s distinct #98221 regression/scope work attributable too. Nobody's contribution needs to be erased for this convergence to land.

### Other side of the shape — keep the dynamic `execute_code` bypass explicitly open

The scope note is right to distinguish this from #98221, so I am **not** adding it as a third blocker for the narrow terminal incident. But it is important not to treat the `tools/code_execution_tool.py` hunk as closure of the larger class. Its hard block still scans source text with `contains_gateway_lifecycle_command(code)`. Runtime construction such as building `"restart"` from pieces can avoid that classifier, and the eventual `hermes gateway restart|stop|uninstall` sink in `hermes_cli/gateway.py` still checks `_is_supervised_gateway_process()` from the CLI child process, which is not the gateway PID owner. The existing `_is_pid_ancestor_of_current_process()` helper shows the sink already has the topology vocabulary needed for a future closure. Please keep that as a distinct follow-up/owner rather than allowing merge of this PR to imply that the execute-code/process-topology class is closed.

### Interlock / landing order

- #93267 is the merged canonical predecessor for the inherited-env false-block side of the lifecycle guard and already preserves #92196/`nbxuhk` + #92633/`DavidMetcalfe` lineage.
- #95954/`giofxjitas` is the direct closed-unmerged predecessor for this direct-spawn identity/supervision split; #98244 should supersede/salvage it with credit, not flatten it.
- Open #98230 also changes `tools/terminal_tool.py`, but in `_rewrite_compound_background`; it is complementary rather than a semantic dependency. Whichever lands second should rebase and rerun exact-head checks rather than infer compatibility from disjoint intent.
- Current `main` has moved well past the upstream commit merged into this head, but I checked the four changed paths and found no current-main commits to them after that integration point. So I do not see a hidden semantic collision from main drift today; after the repairs above, recompose on current main and require fresh exact-head CI.

This is a good, narrow recovery of a real Windows control-plane outage, and the green Windows lane is meaningful evidence. Fixing the one remaining fail-open identity state and preserving the predecessor's authorship would make the narrow #98221 closure substantially stronger without dragging the separate process-topology project into this PR.

chelsealong:
Addressed the P1 on identity ambiguity: `_is_gateway_process()` treated any PID/lock read failure (transient I/O error, malformed metadata) as "definitely not the gateway," so an unproven identity fell through to the allow case on the exact destructive-command path this PR guards. Added `_is_gateway_process_or_unknown()` (built on `gateway.status.get_running_pid_identity_strict()`) and switched both `terminal_tool.py` and `code_execution_tool.py` to it — unproven identity now refuses the command instead of letting it run. `_is_supervised_gateway_process()` (systemd-scope isolation) is untouched. Added regressions for an unreadable identity read and for malformed-but-active identity metadata, plus retargeted the existing guard tests; confirmed all new/retargeted assertions fail against the prior code and pass with the fix. `tests/hermes_cli/test_gateway_restart_loop.py` (301), `tests/tools/test_process_registry.py`+`test_code_execution*.py`+`test_execute_code_approval_cluster.py`+`test_approval.py` (297 passed, 4 skipped), and `ruff check` on the four changed files all pass.

PR #87891, opened 2026-08-16
fix(auth): close Anthropic OAuth CSRF gap, cross-process refresh race, and API-key shadowing

JoaoMarcos44:
## Summary

This PR closes the Anthropic OAuth review gaps without keeping a dashboard implementation of Anthropic subscription OAuth.

- Removes the dashboard Anthropic PKCE flow (`start`, `submit`, and credential-save path).
- Exposes Anthropic in the dashboard as `flow: "external"`; dashboard attempts to start or submit that flow are rejected before creating or persisting an OAuth session.
- Keeps the interactive terminal path (`hermes auth add anthropic`) explicitly outside this dashboard-removal scope.
- Serializes Anthropic refreshes through the credential-store lock.
- Adds a lock keyed to the shared Claude Code credential source, not only the current profile's `auth.json`.
- Applies the same shared-source lock to the direct Claude Code resolver path.
- Preserves refreshed Hermes PKCE singleton state across `load_pool()` reloads without creating a duplicate singleton for `manual:hermes_pkce` pool entries.
- Makes OAuth UI requests follow the selected management profile and rejects cross-profile session polling/cancellation.
- Keeps the Windows lock-file initialization write fail-safe under contention.

## Review follow-up disposition

- CSRF / PKCE leakage and dashboard API-key shadowing: the vulnerable dashboard surface was removed rather than retained with a partial patch.
- Duplicate dashboard identity and cancellation races: the dashboard Anthropic flow and its `manual:dashboard_pkce` creation path no longer exist.
- Claude Code cross-profile race: a shared-source lock and an independent-process regression cover the actual ownership boundary.
- Hermes PKCE reload regression: the exact singleton source is written through after refresh; pool-owned `manual:hermes_pkce` entries remain pool-only.
- OAuth profile scoping: frontend requests carry `?profile=...`; backend session endpoints reject a mismatched profile.
- Windows coverage: the lock stress tests are marked `windows_only` and run against native `msvcrt` on Windows.

### Local validation on Windows

- Focused Anthropic/auth suite: 14 active tests passed, 7 expected macOS-only skips.
- The focused concurrency/persistence suite passed in 3 independent runs with no retry masking.
- `tests/hermes_cli/test_web_oauth_dispatch.py`: 17 passed.
- `web/src/lib/api.test.ts`: 8 passed.
- `npm run typecheck`: passed.
- `npm run lint`: passed with 0 errors and 26 pre-existing warnings.
- Ruff on changed Python files: passed.
[...]

Issue #92606, opened 2026-08-23
Anthropic OAuth (hermes_pkce/dashboard_pkce): stale credential file overwrites rotated token and two pool rows share one single-use refresh token -> 401 revoked -> empty pool


sky0eyes:
## Summary

Anthropic OAuth logins made through Hermes' own PKCE flows (`hermes_pkce`, dashboard `manual:dashboard_pkce`) drift back to a stale, already-rotated refresh token, mutually revoke each other, and end with an empty credential pool. The user-visible sequence is:

1. Mid-session: `Provider error — HTTP 401: OAuth access token has been revoked.`
2. Opening any **new chat** afterwards: `agent init failed: No Anthropic credentials found. Set ANTHROPIC_TOKEN or ANTHROPIC_API_KEY, run 'claude setup-token', or authenticate with 'claude /login'.`

Re-running the OAuth login fixes it until the next rotation, then it recurs.

Environment: Hermes Agent v0.20.5 (2026.8.19), upstream `530028c2`, install method `git`, Windows 11, Python 3.11.16, desktop app, provider `anthropic` / `claude-opus-5`, single profile (`default`).

### Gap A — refresh is never written back for non-`claude_code` sources

`agent/credential_pool.py::_refresh_entry_impl()`, anthropic branch:

```python
refreshed = refresh_anthropic_oauth_pure(
    entry.refresh_token,
    use_json=entry.source.endswith("hermes_pkce"),
)
updated = replace(entry, access_token=..., refresh_token=..., expires_at_ms=...)
# Keep ~/.claude/.credentials.json in sync ...
if entry.source == "claude_code":          # <-- only this source persists
    _write_claude_code_credentials(...)
```

`hermes_pkce` and `manual:dashboard_pkce` entries are refreshed in memory, but `~/.hermes/.anthropic_oauth.json` — the file those sources are seeded *from* — is never updated. Note the code even branches on `entry.source.endswith("hermes_pkce")` two lines above, so that source is explicitly expected here; only its write-back is missing.

### Gap B — the stale file overwrites the fresher pool entry on next load

`_seed_from_singletons()` re-reads `.anthropic_oauth.json` on every pool load and feeds it to `_upsert_entry()` (`agent/credential_pool.py:2435`), which copies any differing field onto the existing entry with no freshness comparison:

```python
for key, value in payload.items():
    if key in {"id", "priority"} or value is None:
        continue
    if key in _field_names:
        if getattr(existing, key) != value:
[...]

PR #98245, opened 2026-08-30
Anthropic OAuth: shared-credential refreshes stop burning single-use tokens; dashboard PKCE removed (salvage #87891)

### LINKED ISSUE #87887: Anthropic dashboard OAuth (PKCE) leaks code_verifier as state and never validates state on callback
## Summary

The web dashboard's Anthropic PKCE login flow (`hermes_cli/web_server.py::_start_anthropic_pkce` / `_submit_anthropic_pkce`) reintroduces a vulnerability class that was already fixed for the CLI flow (`agent/anthropic_adapter.py::run_hermes_oauth_login_pure`, PR #10699 / issue #10693, see `tests/agent/test_anthropic_oauth_pkce.py`).

## Bug 1 — PKCE `code_verifier` sent as OAuth `state`

```python
# hermes_cli/web_server.py::_start_anthropic_pkce
sess["verifier"] = verifier
sess["state"] = verifier  # Anthropic round-trips verifier as state
params = {
    ...
    "code_challenge": challenge,
    "code_challenge_method": "S256",
    "state": verifier,
}
```

The PKCE `code_verifier` (which RFC 7636 §7.2 requires to stay confidential until the token exchange) is placed verbatim in the `state` query parameter of the authorization URL handed to the browser — leaking it via browser history, `Referer` headers, and Anthropic's own access logs.

## Bug 2 — callback never validates `state`

```python
# hermes_cli/web_server.py::_submit_anthropic_pkce
state_from_callback = parts[1] if len(parts) > 1 else ""
exchange_data = json.dumps({
    ...
    "state": state_from_callback or sess["state"],   # never compared to sess["state"]
    ...
}).encode()
```

The endpoint accepts any (or no) `state` value — there is no `if state_from_callback != sess["state"]: reject` check, unlike the CLI flow's `received_state != oauth_state` guard. This removes the CSRF protection RFC 6749 §10.12 requires.

## Reproduction

New regression tests (all fail against the current code, proving the bugs are real):

```bash
python -m pytest tests/hermes_cli/test_anthropic_dashboard_pkce_csrf.py -v
```

## Suggested fix

- Generate an independent `state` via `secrets.token_urlsafe(32)` in `_start_anthropic_pkce`, mirroring the already-fixed CLI flow.
- Reject the exchange in `_submit_anthropic_pkce` when `state_from_callback` is missing or does not match `sess["state"]`.

### LINKED ISSUE #87888: Anthropic OAuth refresh not protected by cross-process lock; stale ANTHROPIC_API_KEY silently shadows OAuth login
## Summary

Two related root causes make "log in with Claude Pro/Max via OAuth" unreliable in practice, even though the login step itself succeeds.

## Bug 3 — `CredentialPool._refresh_entry()` excludes anthropic from the cross-process lock

```python
# agent/credential_pool.py
if self.provider in ("openai-codex", "xai-oauth"):
    with _auth_store_lock(timeout_seconds=self._single_use_refresh_lock_timeout()):
        ...
return self._refresh_entry_impl(entry, force=force)   # anthropic falls through here, unlocked
```

Anthropic's OAuth refresh tokens are single-use too (`agent/anthropic_adapter.py::_refresh_oauth_token`'s own docstring: "a successful refresh rotates the pair and invalidates the old refresh token"), yet `"anthropic"` is missing from the locked-provider tuple that Codex/xAI already get.

The only failure-recovery path, `_sync_anthropic_entry_from_credentials_file`, is hard-scoped to `entry.source == "claude_code"`:

```python
if self.provider != "anthropic" or entry.source != "claude_code":
    return entry
```

So credentials from Hermes's own PKCE login (`hermes_pkce`) or the dashboard (`manual:dashboard_pkce`) have **no recovery at all**. When two Hermes processes race to refresh the same stale token (fleet workers, cron + interactive session, etc.), the loser is marked `STATUS_EXHAUSTED` and prompts for re-auth even though a valid token exists on disk from the winner.

## Bug 4 — dashboard OAuth save never clears a stale `ANTHROPIC_API_KEY`

`resolve_anthropic_token()` deliberately prioritizes an explicit `ANTHROPIC_API_KEY` over the credential-pool OAuth entry:

```python
# 3. Regular API key. An explicit user-configured key must not be shadowed
# by auto-discovered Claude Code or credential-pool OAuth credentials.
api_key = _getenv("ANTHROPIC_API_KEY").strip()
if api_key:
    return api_key
```

The CLI OAuth flow (`save_anthropic_oauth_token` in `hermes_cli/config.py`) clears `ANTHROPIC_API_KEY` when saving a fresh OAuth token, keeping the two slots mutually exclusive. The **dashboard** flow (`_save_anthropic_oauth_creds` in `hermes_cli/web_server.py`) never did this. Result: if `ANTHROPIC_API_KEY` was ever set (leftover from an earlier setup, auto-detection, etc.), logging in via the dashboard with a Claude Pro/Max subscription silently keeps billing pay-per-token against the old key forever — this is the concrete symptom reported by a user ("I set up OAuth and select models, but it still uses the API key").

## Reproduction

```bash
python -m pytest tests/agent/test_credential_pool_anthropic_refresh_race.py tests/agent/test_anthropic_oauth_stress.py -v
```

## Suggested fix

- Add `"anthropic"` to the locked-provider tuple in `_refresh_entry`, and add a provider-agnostic pool-store sync (mirroring `_sync_xai_oauth_entry_from_pool_store`) that works for every Anthropic source, not just `claude_code`.
- Clear `ANTHROPIC_API_KEY` in `_save_anthropic_oauth_creds` the same way the CLI flow already does.

teknium1:
## Summary

Anthropic subscription OAuth no longer races itself across profiles or lies about failed token rotations — salvage of #87891 by @JoaoMarcos44, rebased onto current main with the three remaining review blockers fixed on top.

Root cause family: `claude_code` credentials live in ONE shared `~/.claude/.credentials.json` that every Hermes profile resolves to (confirmed by a six-profile fleet report on the PR), and its refresh tokens are single-use — so concurrent refreshes, lost commits, and the dashboard's parallel PKCE flow could each burn the shared credential family into `invalid_grant`.

## Changes

**Salvaged from #87891 (authorship preserved, 7 commits):**
- `hermes_cli/web_server.py`, `web/src/lib/api.ts`: dashboard Anthropic PKCE flow removed; Anthropic exposed as `flow: "external"` (terminal-only `hermes auth add anthropic`); OAuth session endpoints reject cross-profile polling/cancellation
- `agent/credential_pool.py`: path-keyed cross-process lock on the shared Claude credential source; refresh commit is transactional (fail-closed quarantine on lost writes); borrowed `claude_code` rows no longer treated as token authority; empty OAuth rows unleaseable
- `agent/anthropic_adapter.py` split into four modules (`anthropic_credentials.py`, `anthropic_endpoints.py`, `anthropic_message_convert.py` + slim adapter)

**Follow-up fixes (ours):**
- **Cross-process spent-rotation verdict (last open P1):** the consumed-but-uncommitted verdict was a process-local dict; a fresh interpreter in another process could still lease the stale pair or re-POST the spent refresh token. Now persisted as non-secret one-way fingerprints in a sidecar registry next to the shared singleton source, written under the same path-keyed lock, and consulted before every lease/refresh (pool resolver, pool refresh, direct resolver). Two-process regression test included (A rotates + loses commit; B in a fresh interpreter must fail closed) plus a no-verdict control.
- **Hermetic guard repair (the exact-head CI red):** the module split disarmed `tests/conftest.py`'s macOS Keychain guard (patched a dead re-export) and the setup-token stdin guard (scanned the old file). Both now target the implementation owner and fail loudly if the seam moves again.
- Dropped the two triage-notes markdown files (`ANTHROPIC_ISSUES.md`, `RELATORIO_ANTHROPIC_OAUTH_BUGS.md`) from the tree.

## Validation

| Suite | Result |
|---|---|
| spent-rotation verdict + persist-failure + borrowed-row + hermetic guards + stdin guard | 34 passed |
| full salvaged surface (oauth stress, keychain, pool, refresh race, writethrough, adapter, aux client, web dispatch, lock concurrency) | 376 passed, 9 skipped |
| ruff on changed files | clean |
| two-process witness (fresh interpreter, shared file only) | B sees A's verdict, leases nothing, POSTs nothing |

Credit: @JoaoMarcos44 (all substantive work, 4 review rounds), @andrexibiza (exact-head reviews that shaped the authority boundary), @SilverNine (six-profile fleet evidence; closed their duplicate #95434 in favor of this branch).

Closes #87891. Fixes #87887. Fixes #87888.

## Infographic

Anthropic OAuth: one shared credential, many processes


alt-glitch:

Related: this is a salvage of #87891 and addresses the Anthropic OAuth refresh family tracked by #87888 and #92606.

Issue #97810, opened 2026-08-29
[Bug]: desktop Tasks panel replaces the list on merge:true todo writes


Adolanium:
### Bug Description

The desktop composer Tasks panel treats every live `todo` tool.start as a full list replace. A `merge: true` write with one updated item becomes `Tasks 1/1`, or disappears if the patch has id and status but no content. The running spinner also only appears when the group is collapsed, so an expanded `Tasks 0/5` list looks frozen until the final complete result lands with every item done.

### Steps to Reproduce

1. Open the desktop app and start a turn that creates five todo items, all pending.
2. Have the agent call `todo` with `merge: true` and one item marked completed or in_progress (id plus status is enough, content optional).
3. Watch the Tasks panel on `tool.start`, before `tool.complete`.

### Expected Behavior

The existing five-item list stays. The patched row updates. The in-progress spinner shows on the header while the group is expanded.

### Actual Behavior

`tool.start` parses `args.todos` as a full list. Items without content are dropped. The panel jumps to one row, or empties, until `tool.complete` ships the full store. The header spinner is hidden while the group is expanded, so `Tasks 0/5` looks stuck.

### Affected Component

- Other

### Messaging Platform (if gateway-related)

N/A (desktop composer)

### Debug Report

```shell
N/A. Found by watching the desktop Tasks panel during a live turn. Reproduced with unit tests on merge:true start payloads (status-only patch must keep the rest of the list).
```

### Operating System

Windows 11

### Python Version

3.11

### Hermes Version

main aff5125f8 / desktop 0.17.0

### Root Cause Analysis (optional)

`apps/desktop/src/app/session/hooks/use-message-stream/index.ts` called `parseTodos(payload.todos) ?? parseTodos(payload.result) ?? parseTodos(payload.args)` and replaced the store with that. On `tool.start` only `args` exists. `parseTodos` requires content, so a status-only merge patch becomes `[]`. The Python `TodoStore` merges by id. The desktop client did not.

`StatusSection` only rendered `collapsedIndicator` while collapsed. Todo groups default to expanded.

### Proposed Fix (optional)

Prefer the full result when present. On `merge: true` args, patch by id and allow status-only items. Keep replace when merge is off. Show the in-progress spinner next to the header even when the group is expanded.

PR will follow this issue immediately.

### Are you willing to submit a PR for this?

- [x] I'd like to fix this myself and submit a PR

686f6c61:
[...]

PR #98248, opened 2026-08-30
fix(todo): Tasks panel stays live across merges, reconnects, and hidden tool progress (salvage #97815, #97811)

teknium1:
## Summary

The desktop Tasks panel now stays correct across `merge: true` patches, tool-progress display settings, and session resume/reconnect — with **no new DB tables**.

Salvages the schema-free core of #97815 by @itsflownium and incorporates the focused desktop fix from #97811 by @Adolanium (his commit preserved with authorship). The `session_todo_state` table + copy-on-write lineage machinery from the original PR is intentionally dropped: canonical todo tool results already persist in conversation history, so resume paths derive the snapshot from the stored transcript instead of a parallel store.

Root causes fixed:
1. `tool.start` for a `merge: true` todo write was applied as a full replace, so a one-item status patch collapsed the panel to `Tasks 1/1` (or hid it).
2. `tool.complete` is optional display chrome — suppressed tool progress meant the panel silently froze.
3. Resume/reconnect never re-sent todo state; a renderer restart lost the panel entirely.

## Changes

- `tools/todo_tool.py`: TodoStore gains a monotonic in-memory `revision` (bumped only on real change); the tool result returns it; `restore()` adopts a trusted revision without manufacturing a new one
- `tui_gateway/server.py`: dedicated `todo.updated` full-snapshot event emitted on every todo tool completion, bypassing optional tool-progress display settings; session payloads attach the newest live/cached snapshot; `_read_persisted_todo_state()` derives the resume snapshot from the stored transcript (paired-tool-call check mirrors the GHSA-5g4g-6jrg-mw3g hydration guard)
- `tui_gateway/methods_session.py`: every resume/activate response carries `todo_state`
- `run_agent.py`: history hydration is revision-arbitrated — older history can no longer roll back a newer live store
- desktop: `merge: true` patches merge by id (status-only updates keep the rest of the list), per-session revision store rejects stale updates, `todo.updated` + resume snapshots restore the panel, running indicator stays visible while the Tasks section is expanded

## Validation

| Check | Result |
|---|---|
| `tests/tools/test_todo_tool.py` + `tests/tui_gateway/test_todo_state_events.py` | 21/21 pass |
| `tests/run_agent/test_run_agent.py` | 281/281 pass |
| Desktop vitest (4 todo suites) | 37/37 pass |
| Desktop `tsc --noEmit` | clean |
| E2E vs real SessionDB (temp HERMES_HOME) | transcript-derived resume snapshot OK; forged bare tool message rejected; revision monotonicity OK |
| ruff | clean |

Addresses the live-panel half of #19477 (restart persistence of the *panel*; agent-side state was already recoverable from history).

## Credits

- @itsflownium — #97815: revision design, `todo.updated` event, resume attach (salvaged, slimmed)
- @Adolanium — #97811: desktop merge-instead-of-replace fix (commit preserved)

## Infographic

Live task state, no new tables

### DIFF
--- a/apps/desktop/src/app/chat/composer/status-stack/collapsed-indicator.test.tsx
+++ b/apps/desktop/src/app/chat/composer/status-stack/collapsed-indicator.test.tsx
@@ -22,6 +22,22 @@ describe('ComposerStatusStack collapsed todo indicator', () => {
     $todosBySession.set({})
   })
 
+  it('shows a running indicator while the todo group is expanded', () => {
+    $todosBySession.set({
+      'session-1': [{ content: 'Wire the status stack', id: '1', status: 'in_progress' }]
+    })
+
+    render(
+      <MemoryRouter>
+        <ComposerStatusStack queue={null} sessionId="session-1" />
+      </MemoryRouter>
+    )
+
+    expect(screen.getByText('Wire the status stack')).toBeTruthy()
+    expect(screen.getAllByRole('status').length).toBeGreaterThan(0)
+    expect(screen.getByText('Tasks 0/1')).toBeTruthy()
+  })
+
   it('shows a running indicator next to the collapsed todo label', () => {
     $todosBySession.set({
       'session-1': [{ content: 'Wire the status stack', id: '1', status: 'in_progress' }]
--- a/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/tools.ts
+++ b/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/tools.ts
@@ -4,6 +4,7 @@ import { flashPetActivity, setPetActivity } from '@/store/pet'
 import { pruneDelegateFallbackSubagents, upsertSubagent } from '@/store/subagents'
 import { reportMcpToolResult } from '@/store/suggestion-providers/repair'
 import { invalidateSkillSuggestionIndex } from '@/store/suggestion-providers/skill'
+import { restoreSessionTodosFromSnapshot } from '@/store/todos'
 import { recordToolDiff } from '@/store/tool-diffs'
 import { setSessionDraftingTool } from '@/store/tool-drafting'
 import { notifyWorkspaceChanged, toolChangedPath, toolMayMutateFiles } from '@/store/workspace-events'
@@ -16,6 +17,14 @@ import type { GatewayEventContext } from './types'
 export function handleToolEvent(ctx: GatewayEventContext): boolean {
   const { deps, event, payload, sessionId, isActiveEvent, occurredAt } = ctx
   const { flushQueuedDeltas, nativeSubagentSessionsRef, sessionInterrupted, updateSessionState, upsertToolCall } = deps
+
+  if (event.type === 'todo.updated') {
+    if (sessionId && !sessionInterrupted(sessionId)) {
+      restoreSessionTodosFromSnapshot(sessionId, payload, true)
+    }
+
+    return true
+  }
 
   if (event.type === 'tool.generating') {
     // Announced while the model is still emitting the call's JSON, so it
--- a/apps/desktop/src/app/session/hooks/use-message-stream/index.ts
+++ b/apps/desktop/src/app/session/hooks/use-message-stream/index.ts
@@ -23,7 +23,7 @@ import {
   generatedImageEchoSources,
   stripGeneratedImageEchoes
 } from '@/lib/generated-images'
-import { nextTodosFromToolEvent } from '@/lib/todos'
+import { nextTodosFromToolEvent, parseTodoRevision } from '@/lib/todos'
 import { dispatchNativeNotification } from '@/store/native-notifications'
 import { isDiskFullErrorMessage, notifyError } from '@/store/notifications'
 import { broadcastSessionsChanged } from '@/store/session-sync'
@@ -465,7 +465,7 @@ export function useMessageStream({
         const todos = nextTodosFromToolEvent($todosBySession.get()[sessionId] ?? [], payload)
 
         if (todos) {
-          setSessionTodos(sessionId, todos)
+          setSessionTodos(sessionId, todos, parseTodoRevision(payload))
         }
       }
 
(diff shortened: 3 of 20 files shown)


100yenadmin:
## Field report: Desktop tasks invisible with tool progress disabled on a pre-fix installation

Adding a detailed real-world report for the exact hidden-tool-progress failure addressed by this PR. **This is confirmation from an affected older installation, not a claim that the bug reproduces on current upstream `main`, and not a request to reopen this PR.** The upstream check during diagnosis located this merged fix, so this report belongs here rather than in a duplicate issue.

### User-visible impact

The user could no longer see plan/task updates in Hermes Desktop and reasonably suspected that planning had stopped working or the tool had disappeared. In fact, the agent was successfully maintaining the checklist; the UI did not show it even during the active turn. Hiding verbose tool activity had also hidden meaningful task state.

The native tool exposed in the affected session is `todo`. The user referred to `update_plan` as well; we did not establish that an upstream tool with that name was removed or renamed.

### Environment and version boundary

- macOS 26.6.2, arm64; Python 3.11.15.
- Packaged Electron Hermes Desktop, with a git-installed backend and locally carried commits.
- CLI version output at filing: `Hermes Agent v0.20.6 (2026.8.27) · upstream 42da1f2c · local 3a215731 (+169 carried commits)`.
- Initial source inspection used local HEAD `98198a12bea152c055fbc4587d1016bcc5e5b618`. Filing-time HEAD was `3a21573132246cd6e961f6419a6c3fd0f21af6ce`; `git diff` between those commits showed no differences in `tui_gateway/server.py` or `apps/desktop/src/store/todos.ts`.
- Relevant configuration: `display.tool_progress: false`.
- Exact correspondence between the running packaged renderer and inspected source was **not established**. This was not a clean current-main installation or an end-to-end acceptance run of the merged fix.

### Observed behavior

1. `todo()` was available and returned successfully.
2. A real three-item diagnostic checklist was created, then updated with `merge: true`. Writes and returned states succeeded; corresponding tool results were recorded in the session database.
3. Two live Desktop screenshots taken after successful updates, while the investigation was still running, showed no checklist/plan. Almost no ordinary tool cards were visible either.
4. Targeted inspection of other recent Desktop sessions also found successful todo calls, ruling out “the agent simply never uses the task tool” as an explanation for this installation.
5. There was no todo execution error or exception associated with the missing panel.

No private screenshots, conversation transcripts, database rows, or raw configuration dumps are attached.

### Minimal reproduction recipe for an affected build

1. In an isolated profile, set `display.tool_progress: false` before starting the Desktop backend/session. Ensure the effective session tool-progress mode is `off`; an environment or session override can otherwise change the test.
2. Ask the agent to create a `todo` list with one `in_progress` and at least one `pending` item, then continue a bounded multi-step task so the turn remains active long enough to inspect the composer.
3. Verify the tool returns the list, then update one item using `merge: true` and verify the full resulting list.
4. Observe the task/status stack above the composer during the turn, not only after the final response.
5. Compare with an isolated progress-enabled session and with a backend/renderer pair containing this PR.

Steps 1–4 describe the diagnosed configuration and live symptom. The progress-on A/B run and upgraded-build comparison in step 5 were **not performed** in this investigation.

### Source-level cause in the affected version

In the inspected `tui_gateway/server.py`:

- `_load_tool_progress_mode()` at lines 5853–5863 maps configuration boolean `false` to `off`.
- `_tool_progress_enabled()` at lines 6049–6050 returns false for that mode.
- `_tool_lifecycle_required_for_ui()` at lines 6053–6059 exempts `clarify` and `setup_mcp`, but not `todo`.
- `_on_tool_complete()` at lines 7689–7690 only emits `tool.complete` if progress is enabled, an inline diff exists, or the tool is exempt.

The older renderer derives its task state from the todo tool lifecycle events. Consequently a successful canonical todo result can be saved and returned to the agent without producing the live UI update. The missing panel during the active turn is distinct from normal turn-end cleanup.

This matches the mechanism explicitly fixed by [393af4a310e6e283901b62ed841d0eb322a1df79](https://github.com/NousResearch/hermes-agent/commit/393af4a310e6e283901b62ed841d0eb322a1df79): dedicated `todo.updated` snapshots independent of optional tool progress, compatibility completion events, and authoritative resume/activation snapshots. GitHub reports this PR merged on August 30, 2026, with merge commit `c0875ba503ff1d03978831efb159443764fc4db4`.

### Separate UX concern: plan visibility after a turn

The older store explicitly says the inline transcript panel was removed in favor of the composer status stack. It clears unfinished lists at turn end and lets finished lists linger for four seconds. That behavior amplifies the impression that plans disappeared, but should **not** be conflated with the live-event suppression bug or blindly reversed: stale `in_progress` text must not imply that execution is still running.

Existing related work:

- #75665 — open proposal for a persistent, read-only Current Plan surface after turns. This is the appropriate place for durable plan visibility, separate from live execution status.
- #100652 — idle historical sessions briefly replay completed task lists; distinct from this report's missing active-turn panel.
- #97810 — merge-patch list replacement, another separate symptom addressed by this salvage PR.

Current upstream renderer source inspected at `549e6aab38c493fc0d3fb0494735d27ce95f0c78` includes revision arbitration and restoration of active snapshots when the backend reports `running === true`. The older installation's hydration behavior should therefore not be generalized to current upstream.

### Suggested installation/acceptance checks

For an upgrade or packaging validation of this fix:

- [ ] With tool progress off, a successful todo write visibly updates the active Desktop Tasks panel while unrelated tool chatter remains hidden.
- [ ] A status-only `merge: true` update preserves the other rows and updates the intended item.
- [ ] Resume/reconnect during an actually running turn restores the authoritative checklist.
- [ ] A stale snapshot cannot overwrite a newer revision; an explicit empty snapshot clears the list.
- [ ] Completed/cancelled and idle-history behavior does not resurrect stale work as live execution.
- [ ] Validate the installed backend and packaged renderer together; source presence alone is not installed-product proof.

The relevant upstream regression tests are `tests/tui_gateway/test_todo_state_events.py` and the Desktop todo store/message-stream suites. Their presence and the merged patch were inspected; **no tests were run and no runtime configuration, code, or installation was changed as part of this report**.

The remaining local action is a separately approved update/packaging verification, not a new upstream implementation of the same fix.

PR #98256, opened 2026-08-30
fix(desktop): clear a removed connection's cookie jar on removal

### LINKED ISSUE #98242: Desktop: re-adding a remote gateway flashes 'Signed in' and skips login — connection still fails
## Bug description

On the macOS desktop app, re-adding a remote gateway connection never shows the login page. The app flashes through a "Signed in" state and the connection test still fails. The gateway backend is healthy and the credentials are valid (verified with curl), so the failure is in the app's auth-state handling, not the server.

## Steps to reproduce

1. Add a remote gateway (e.g. `http://<lan-host>.local:9119`) in the desktop app and complete the login once.
2. Remove the connection, or let the session lapse, then re-add the same URL.
3. Expected: the login page opens again.
4. Actual: the app flashes "Signed in" without ever showing the login page, and the Test probe / connection still fails.

## Evidence

- Gateway `/api/status` → `auth_required: true`, `auth_providers: ["basic"]`, `auth_flows: ["cookie", "native_pkce"]` (reachable, healthy).
- `POST /auth/password-login` with the correct credentials → `{"ok":true,"next":"/"}` (credentials valid).
- `GET /login` → HTTP 200 (login page exists).
- Deleting the app's `connections.json` / `connection.json` and the native token store does not restore the login page — the flash-through persists.
- Desktop log shows no browser-launch or OAuth errors — the login flow is being skipped, not failing.

## Expected behavior

Re-adding a connection must always present the login flow (or a clear re-auth affordance). The app must not report "Signed in" when no valid session exists.

## Environment

- macOS desktop app (self-updated Aug 29, 2026; issue appeared after that update)
- Remote gateway: self-hosted Hermes gateway on LAN, `basic` (username/password) auth

## Related

- #66521 — Desktop Remote gateway shows legacy session-token field, not username/password Sign In
- #61457 — remote gateway session cookie never persists after basic-auth login
- #95609 — native PKCE failure silently downgrades to embedded cookie login

chelsealong:
## What does this PR do?

Fixes the desktop app flashing "Signed in" and skipping the login page when
a remote gateway connection is removed and then re-added at the same URL,
while the actual connection still fails.

Root cause: `hermes:connections:remove` tore down the removed connection's
pooled backends and SSH tunnels, but never cleared its Electron session
cookies. A primary v2 registry connection rides the shared legacy oauth
partition (`persist:hermes-remote-oauth`), so removing it and later
re-registering the same URL as the new primary inherits whatever cookies
were left behind in that shared jar — often a session that is already
invalid server-side (revoked, expired, or from before a credential change).
`hasLiveOauthSession()`'s cookie-presence check then reports "connected"
immediately, so the settings/reconnect UI shows "Signed in" and never opens
the login window, while the authoritative check (the ws-ticket mint at
actual connect time) still fails — matching the reported symptom exactly,
including that wiping `connections.json` and the native token store did
not help, since neither of those touches Electron's separate cookie-jar
storage for the partition.

## Related Issue

Fixes #98242

## Type of Change

## Changes Made

- `apps/desktop/electron/oauth-partition.ts` — add `cookieJarToClearOnRemoval(id, registry)`,
  a pure decision of which URL's cookie jar must be cleared when a connection
  is removed (covers both the shared legacy jar for a primary connection and
  the per-connection jar for a secondary), or `null` when the connection
  never used cookie auth.
- `apps/desktop/electron/main.ts` — `hermes:connections:remove` now clears
  the removed connection's cookie jar (via the existing `clearOauthSession`)
  *before* writing the registry without that entry, since partition
  resolution depends on the entry still being present.
- `apps/desktop/electron/oauth-partition.test.ts` — regression tests for
  `cookieJarToClearOnRemoval` covering a cookie-auth (oauth/basic) remote, a
  token-auth remote (should not clear), the local connection, and an unknown
  id.

## How to Test

1. `npm ci --workspace apps/desktop --include-workspace-root`
2. `npx vitest run electron/oauth-partition.test.ts --project electron` (from `apps/desktop`)
3. `npx tsc -p apps/desktop/tsconfig.electron.json --noEmit`
4. `npx eslint apps/desktop/electron/oauth-partition.ts apps/desktop/electron/oauth-partition.test.ts apps/desktop/electron/main.ts`

### Test output

```
 Test Files  1 passed (1)
      Tests  15 passed (15)
```

Confirmed the 4 new tests fail without the fix (temporarily reverted
`oauth-partition.ts`/`main.ts` to HEAD while keeping the new test file):

```
FAIL electron/oauth-partition.test.ts > cookieJarToClearOnRemoval (#98242 stale session on re-add) > names the URL of a removed cookie-auth (oauth/basic) connection
TypeError: cookieJarToClearOnRemoval is not a function
 Test Files  1 failed (1)
      Tests  4 failed | 11 passed (15)
```

Also ran the full electron vitest project (`npx vitest run --project electron`
from `apps/desktop`): **1958 passed, 6 skipped, 0 failed** — no regressions.

## Checklist

- [x] I've added tests for my change, and verified they fail without the fix
- [x] Ran typecheck (`tsc -p tsconfig.electron.json --noEmit`) and lint (`eslint`) on touched files — clean

## AI assistance disclosure

This fix (investigation, patch, and tests) was authored with the assistance
of an AI coding agent (Claude), with the resulting diff reviewed before
opening this PR.

### DIFF
--- a/apps/desktop/electron/main.ts
+++ b/apps/desktop/electron/main.ts
@@ -266,7 +266,7 @@ import {
 import { runNativeLogin } from './native-oauth-login'
 import { loadNativeTokenSet, type NativeTokenStoreIo, persistNativeTokenSet } from './native-token-store'
 import { serializeJsonBody, setJsonRequestHeaders } from './oauth-net-request'
-import { LEGACY_OAUTH_PARTITION, resolveOauthPartition } from './oauth-partition'
+import { cookieJarToClearOnRemoval, LEGACY_OAUTH_PARTITION, resolveOauthPartition } from './oauth-partition'
 import { createParentStartMarkerResolver, parentWatchdogEnv } from './parent-process-identity'
 import { registerPetOverlayIpc } from './pet-overlay-ipc'
 import {
@@ -14705,7 +14705,21 @@ ipcMain.handle('hermes:connections:save', async (_event, payload) => {
 ipcMain.handle('hermes:connections:remove', async (_event, id) => {
   const key = String(id || '')
   managedConnectionUpdateGate.assertCanMutate(key)
-  const registry = removeConnection(readDesktopConnectionsRegistry(), key)
+  const before = readDesktopConnectionsRegistry()
+  const cookieUrlToClear = cookieJarToClearOnRemoval(key, before)
+
+  // Clear the removed connection's cookie jar before its registry entry is
+  // gone. clearOauthSession() re-resolves the partition from the on-disk
+  // registry, so this must run before writeDesktopConnectionsRegistry() below
+  // removes the entry that resolution depends on. Otherwise a later re-add of
+  // the same URL inherits whatever cookies were left behind — often a
+  // long-since-invalidated session — and the UI reports "Signed in" from the
+  // stale jar without the login window ever running (#98242).
+  if (cookieUrlToClear) {
+    await clearOauthSession(cookieUrlToClear)
+  }
+
+  const registry = removeConnection(before, key)
   writeDesktopConnectionsRegistry(registry)
   // Tear down anything the removed connection still had running: pooled
   // backends under its composite keys and any ssh tunnel scopes it owned.
--- a/apps/desktop/electron/oauth-partition.test.ts
+++ b/apps/desktop/electron/oauth-partition.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest'
 
-import { LEGACY_OAUTH_PARTITION, resolveOauthPartition } from './oauth-partition'
+import { cookieJarToClearOnRemoval, LEGACY_OAUTH_PARTITION, resolveOauthPartition } from './oauth-partition'
 
 // #92183 — two basic-auth (cookie-flow) gateways registered in the v2
 // connections registry must not share one cookie jar. Chromium cookie jars
@@ -140,5 +140,36 @@ describe('resolveOauthPartition (#92183 per-connection cookie jars)', () => {
     const got = resolveOauthPartition('https://gw-a.example.com', { registry: reg })
 
     expect(got).toContain('alpha')
+  })
+})
+
+// #98242 — re-adding a removed remote gateway flashed "Signed in" and skipped
+// the login page. Removing a connection never cleared its cookie jar, so
+// re-registering the same URL (most commonly as the new primary, which rides
+// the shared legacy jar) inherited stale cookies from the old, already-dead
+// session and read as signed in without ever running the login flow.
+describe('cookieJarToClearOnRemoval (#98242 stale session on re-add)', () => {
+  it('names the URL of a removed cookie-auth (oauth/basic) connection', () => {
+    const reg = registry('conn-a', [remote('conn-a', 'https://gw.example.com')])
+
+    expect(cookieJarToClearOnRemoval('conn-a', reg)).toBe('https://gw.example.com')
+  })
+
+  it('returns null for a token-auth connection (never rode a cookie jar)', () => {
+    const reg = registry('local', [remote('conn-a', 'https://gw.example.com', { authMode: 'token' })])
+
+    expect(cookieJarToClearOnRemoval('conn-a', reg)).toBeNull()
+  })
+
+  it('returns null for the local connection', () => {
+    const reg = registry('local', [{ id: 'local', kind: 'local' }])
+
+    expect(cookieJarToClearOnRemoval('local', reg)).toBeNull()
+  })
+
+  it('returns null for an id that is not in the registry', () => {
+    const reg = registry('local', [remote('conn-a', 'https://gw.example.com')])
+
+    expect(cookieJarToClearOnRemoval('conn-b', reg)).toBeNull()
   })
 })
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** Fixes re-add illusion of "Signed in" without login (#98242) by clearing the doomed connection's cookie jar before its registry entry is removed (partition re-resolution depends on that entry).

**What it does**
- `apps/desktop/electron/oauth-partition.ts:103` adds `cookieJarToClearOnRemoval(id, registry)` — returns `entry.url` only if `kind==='remote' && authMode==='oauth'`, else `null`; requires registry snapshot before removal (shared legacy `persist:hermes-remote-oauth` vs per-connection scope depends on entry still present).
- `apps/desktop/electron/main.ts:266` imports it, and `hermes:connections:remove` computes `cookieUrlToClear = cookieJarToClearOnRemoval(key, before)` and `await clearOauthSession(cookieUrlToClear)` BEFORE `removeConnection(before,key)` + `writeDesktopConnectionsRegistry`.
- Adds `oauth-partition.test.ts:59` covering cookie-auth URL returned, token-auth/local/unknown → null.

**Non-blocking notes**
- `clearOauthSession` is awaited prior to registry write — correct ordering per comment; confirm `clearOauthSession` is idempotent if URL's jar was already empty (it is — deletes cookies by URL/partition, no throw on miss).
- `authMode==='oauth'` strict check means `basic` mode not cleared via this helper — legacy basic auth rides cookie jar too? Original bug description says primary v2 rides shared legacy oauth partition; check if basic should also clear (separate path or handled elsewhere). Current guard matches filed root cause.
- No change to `primary` handling — stale jar inheritance via re-adding same URL as new primary is now prevented.

liyangbing:
<p>The fix uses the right ordering: it resolves the cookie jar from the pre-removal registry snapshot, clears it, and only then writes the registry without the connection. That addresses the stale-session path where a re-added URL could inherit cookies from a removed primary connection and make the UI claim <code>Signed in</code> before the authoritative connection check.</p>

<h3>Suggested contract</h3>
<ul>
<li>Keep the pre-removal snapshot and the clear-before-write ordering atomic from the handler's point of view; a failed <code>clearOauthSession()</code> should not silently report a completed removal while stale authentication remains.</li>
<li>Align the implementation and contract for cookie-auth modes: the helper currently returns a URL only when <code>kind === remote</code> and <code>authMode === oauth</code>, while the PR description and tests describe “oauth/basic”. If basic-auth remotes also use an Electron cookie jar, add that case; otherwise document that the scope is OAuth-only.</li>
<li>Clear the exact partition selected by the pre-removal registry, including the shared legacy partition for the primary connection and the per-connection partition for a secondary connection. Do not clear another connection's jar when two entries share a URL or when the removed ID is stale.</li>
<li>Preserve the existing teardown and registry-write behavior when no cookie jar applies, and make repeated removal or a missing/invalid registry snapshot safe and idempotent.</li>
</ul>

<h3>Regression matrix</h3>
<ul>
<li>remove a primary OAuth connection, then re-add the same URL as primary: the old cookie jar is empty before the new login check</li>
<li>remove a secondary OAuth connection and verify only its connection-scoped partition is cleared</li>
<li>OAuth/basic/token remotes, local connections, unknown IDs, malformed registry data, and repeated removal</li>
<li>make <code>clearOauthSession()</code> fail and verify the handler's returned error, registry state, and retry behavior are explicit</li>
<li>exercise the full IPC handler with a spy/mock, not only the pure URL helper, to lock in clear-before-registry-write ordering</li>
</ul>

<p>The four focused helper tests and the reported full Electron suite establish the partition-selection cases. The handler-level failure path and the OAuth/basic scope distinction are the remaining high-value checks.</p>

chelsealong:
Thanks for the review. On the oauth/basic scope point: `authMode` only ever takes `'oauth' | 'token'` in this codebase — dashboard basic/password providers are represented as `authMode: 'oauth'` (already noted in the module header for `resolveOauthPartition`), so `cookieJarToClearOnRemoval`'s `authMode === 'oauth'` check already covers both; I've clarified that in its doc comment (095df2f9c).

On the rest of the suggested contract: the ordering/atomicity and primary-vs-secondary partition selection are already covered — `clearOauthSession` is a pre-existing best-effort helper (never throws, same as its other callers) and partition selection is delegated to the already-tested `resolveOauthPartition`. I held off on adding a full IPC-handler-level spy/mock test for `hermes:connections:remove` — main.ts's electron handlers aren't unit-tested at that level elsewhere in this codebase, and mocking that harness didn't seem justified for a best-effort no-throw path; happy to revisit if you feel strongly.

chelsealong:
Looked into the failing "JS & TS checks" job (run 33301376356, job 99230131661). The only failing unit is `apps/desktop :: check:test:ui`, and within it a single test:

```
FAIL src/app/settings/gateway-settings.test.tsx > GatewaySettings > loads the machine-level connection config (no profile scoping)
Error: Test timed out in 15000ms.
Test Files  1 failed | 682 passed (683)
     Tests  1 failed | 6751 passed (6752)
```

That file isn't touched by this PR (only `apps/desktop/electron/{oauth-partition.ts,main.ts,oauth-partition.test.ts}` are). Locally I reproduced the exact CI commands from a clean `npm ci`:

- `check:lint` (tsc x3 + eslint): pass, 0 errors
- `test:ui` (`vitest run --project ui`): 683 files / 6728 tests, **0 failed**, including that specific test both in isolation and inside the full run
- `test:desktop:platforms` (`vitest run --project electron`): 1958 passed, matching the PR description
- `test:desktop:all` (full build + electron-builder package + bundle validation): pass

So this looks like a timing flake under CI's parallel load (one test hit its 15s timeout in the full run), not something introduced by this change. I don't have rerun permissions on this fork's Actions run — a maintainer re-running the job should go green.

PR #86732, opened 2026-08-15
fix(kanban): refuse a review row that no dispatcher can pick up

moonweave:
## What does this PR do?

`request_review` can park a row in `review` with no reviewer and no assignee. Nothing downstream will ever pick that row up.

The dispatcher's review lane selects `status='review' AND claim_lock IS NULL` and then **skips rows with no assignee**. There is no CLI verb and no tool that claims a row already in `review`. So a row that reaches `review` unassigned is not waiting for a reviewer — it is stranded, and it stays stranded until a human notices and assigns it by hand.

`request_review` only recovers a reviewer from a prior `changes_requested` event, which is the re-review path. On a **first** review with neither `reviewer=` nor an inheritable assignee, it succeeded and produced exactly that unreachable row.

Observed 2026-08-14 on a live board: a builder finished at 20:54 and requested review with an empty assignee. The row sat for 12 hours until a person assigned it the next morning. The board looked busy the whole time and nothing was moving.

Refusing at the transition turns a silent stall into an error the caller has to answer.

## Related Issue

No existing issue.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `hermes_cli/kanban_db.py` — `request_review` refuses when a first review has neither an explicit `reviewer` nor an assignee to inherit. Inheriting an assignee still works; naming a reviewer still works. **Only the case where both are absent** is refused.
- `tests/hermes_cli/test_kanban_review_lifecycle.py` — regression test for the refusal, plus coverage that both accepting paths are unaffected.

### Deliberately not included

An earlier draft also blocked an implementer from reviewing their own work. That broke 13 existing tests: a reviewer-less `request_review` keeping the assignee and the dispatcher re-spawning the same profile with the `sdlc-review` skill is the **designed** contract, not an accident. That is a separate policy argument and does not belong in a stall fix.

## How to Test

1. `pytest tests/hermes_cli/test_kanban_review_lifecycle.py -q` → 20 passed.
2. Positive control: revert the guard in `request_review` and re-run → `test_first_review_without_reviewer_or_assignee_is_refused` fails.
3. Manually: `kanban create` a task with no assignee, claim it, `request_review` with no `reviewer` → refused with a message naming the reason.

### Code

- [x] I've read the Contributing Guide
- [x] My commit messages follow Conventional Commits
- [x] I searched for existing PRs to make sure this isn't a duplicate
- [x] My PR contains **only** changes related to this fix
[...]

PR #46485, opened 2026-06-15
fix(kanban): reject reroute self-parent deadlocks

ahmadashfq:
## What does this PR do?

Fixes the permanent deadlock caused when a dispatcher-spawned worker calls `kanban_create` and lists its own `HERMES_KANBAN_TASK` in `parents`.

The original PR proposed inverting that self-edge so the spawning task would wake after the child. Since then, the fleet's live behavior changed: current Hermes rejects self-parent / cycle cases with `ValueError` instead of silently rewriting the dependency graph. The upstream vehicle for #46485 should match that live behavior.

This refresh re-scopes the PR to the reject-based approach:
- `tools/kanban_tools.py`: `kanban_create` now rejects using the current worker task id in `parents`, with an explicit deadlock-oriented error.
- `hermes_cli/kanban_db.py`: `create_task` now reuses `_would_cycle(...)` before inserting parent links, so create-path semantics match `kanban_link` cycle checks.
- tests now model real upstream dependencies instead of encoding the buggy self-parent pattern.

## Related issue / evidence

No upstream issue.

Authoritative reconciliation evidence for the re-scope is in:
- `~/.hermes/audits/hermes-patch-reconciliation-20260707/findings.md` (item 6)
- `~/.hermes/audits/hermes-patch-reconciliation-20260707/execution-checklist.md`

Those artifacts record that the fleet has already been running the reject-with-`ValueError` semantics since the 2026-06-23 safe update, while the old inversion / `wake_after` approach is no longer the live behavior.

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes made

- `tools/kanban_tools.py`
  - reject `parents=[<current worker task>]` with a clear error explaining the reroute deadlock
- `hermes_cli/kanban_db.py`
  - add `_would_cycle(...)` enforcement inside `create_task(...)` parent-link insertion
- `tests/tools/test_kanban_tools.py`
  - update create-path tests to use real upstream dependencies
  - add regression coverage for self-parent rejection
  - update lifecycle expectations so rerouted children wait only on real upstream deps
- `tests/hermes_cli/test_kanban_db.py`
  - add regression coverage that create-time parent linking raises `would create a cycle`

## How to test

Targeted verification run in the refreshed worktree:

```bash
python -m py_compile tools/kanban_tools.py hermes_cli/kanban_db.py tests/tools/test_kanban_tools.py tests/hermes_cli/test_kanban_db.py
[...]

PR #98259, opened 2026-08-30
fix(kanban): break goal-judge deadlock on same-card review requests

### LINKED ISSUE #98160: [Bug]: Goal-mode Kanban task deadlocks on same-card request_review vs reviewer close-out
## Bug description

A goal-mode Kanban implementation task that requires first-class same-card review can enter a circular terminal-action constraint: the goal judge rejects `kanban_request_review` because reviewer close-out does not exist yet, while completion is also rejected because the required review has not happened. The worker can only block for human intervention.

## Reproduction from a real task

Hermes Agent upstream/build: `4209d371` (local carried commit does not touch this lifecycle).

Task `t_8d5ee37e` was created with `--goal` and acceptance explicitly required same-card review. After implementation commit `c21e7b8` and all gates passed, recovery worker `juniorjosh` attempted the required review handoff.

Observed canonical runs/events:

- run 41: `kanban_request_review` could not be accepted; worker blocked with: `Goal-judge rejects same-card request-review until Picasso review plus close-out already exist`
- later attempts repeated the same cycle; run 44 blocked: `Goal-judge circular on request-review vs complete`
- manual recovery required a separate read-only Picasso card `t_a89f0651`, then a human/operator-created zero-duration completion run 45 on the origin task

The card accumulated repeated blocker comments and duplicate recovery attempts even though implementation, gates, and preview were complete.

## Expected behavior

For a goal-mode card whose contract requires same-card review:

1. `kanban_request_review(...)` must be a valid terminal action for the implementation run.
2. The goal judge should evaluate implementation completion for the current phase, not require future reviewer evidence before allowing transition into `review`.
3. Reviewer completion should satisfy the overall goal only after the review run.

## Actual behavior

The judge requires reviewer close-out before permitting the transition that creates the reviewer run, producing a circular dependency. Blocking is the only accepted terminal action.

## Suggested regression test

Create a `--goal` code task whose body requires same-card review, claim as implementer, call `kanban_request_review(reviewer=...)`, and assert:

- implementation run closes with `outcome=review_requested`;
- card enters `review`;
- reviewer can claim it;
- no goal-judge rejection/block is emitted before reviewer execution.

## Workaround used

Create a separate non-goal read-only review card, complete it, then manually close the original task with review metadata. This defeats the intended same-card lifecycle and should not be necessary.

chelsealong:
## What does this PR do?

A goal-mode Kanban card whose own acceptance criteria require same-card
review can deadlock permanently. `kanban_request_review` and
`kanban_complete` both gate goal-mode handoffs on the identical question
put to the auxiliary judge: *"is the whole card (title + body) done,
given this evidence?"* For a card that requires reviewer sign-off before
being "done," that question can never pass from an implementer's summary
alone — the review hasn't happened yet, and it can't happen until
`kanban_request_review` succeeds. The result observed in the issue: the
judge rejects `kanban_request_review` for lacking reviewer-approval
evidence, and rejects `kanban_complete` for the same reason — the worker
can only block for a human, and a manual side-channel (creating a
separate read-only review card, completing it, then hand-closing the
original) is needed to escape.

This scopes the judge question asked at a **review** handoff to
"is the implementation ready to hand off to a reviewer," instructing it
not to withhold `DONE` merely because reviewer/approval evidence is
absent. `kanban_complete` is untouched and keeps asking the original,
unscoped question — a worker still cannot bypass real completion
evidence by routing through `kanban_request_review` first, since the
reviewer's own eventual `kanban_complete` still runs the full check.

**Follow-up fix from review:** the scoping note was originally appended
*after* `title + body`, and `judge_goal` truncates its `goal` argument to
2000 characters before it ever reaches the judge. For a card whose
title+body already runs long — routine, since `task.body` has an 8 KB
budget elsewhere in the codebase — the note (or its operative sentence)
was silently cut off, so `kanban_request_review` would still be judged
against the original unscoped question for exactly the moderately
detailed, explicit-review-required cards this PR targets. Fixed by
reserving a budget for the note (`2000 - len(note)` chars) and truncating
title+body to that budget *before* appending the note, in both duplicate
copies of the gate, so the full note always survives within
`judge_goal`'s own 2000-char limit.

## Related Issue

Fixes #98160

## Type of Change

## Changes Made

- `tools/kanban_tools.py` — `_goal_mode_handoff_rejection` takes a
  `handoff` kwarg (`"complete"` default, `"review"` for
  `kanban_request_review`); when `"review"`, a note is appended to the
  goal text sent to the judge scoping the question to implementation
  readiness rather than full-card completion. Title+body is now
  truncated to a reserved budget first so the note itself can never be
  cut off by `judge_goal`'s 2000-char limit.
- `hermes_cli/kanban.py` — the CLI's independent copy of the same gate
  (used by `hermes kanban request-review` / `/kanban request-review`)
  gets the identical `handoff="review"` scoping and budget reservation,
  so the CLI path and the tool-call path stay consistent.
- `tests/tools/test_kanban_tools.py` — two regression tests:
  - a mock-judge test proving `kanban_request_review` succeeds on
    implementation-only evidence while `kanban_complete` remains gated
    exactly as before;
  - a test with a realistic long card body (>1800 chars title+body) that
    patches the LLM transport underneath the *real* `judge_goal` (not
    `judge_goal` itself, which would bypass its truncation and mask the
    bug) and asserts the review-scoping note is still present in the
    prompt actually sent to the model after truncation.

## How to Test

```
python -m pytest tests/tools/test_kanban_tools.py -q
python -m pytest tests/hermes_cli/test_kanban_review_lifecycle.py tests/hermes_cli/test_kanban_review_lifecycle_complete.py tests/hermes_cli/test_kanban_review_surfaces.py tests/hermes_cli/test_kanban_diagnostics.py tests/hermes_cli/test_kanban_goal_mode.py -q
```

Result: `34 passed` (kanban_tools alone). Combined run across all six
files above: `91 passed, 2 failed` — the 2 failures
(`ModuleNotFoundError: No module named 'fastapi'` and `'acp'`) are
pre-existing missing-dependency errors in this sandbox, unrelated to
this change; confirmed identical on unmodified `HEAD` before this
commit.

**Proof the new truncation-regression test fails without the fix:**
reverted `tools/kanban_tools.py` and `hermes_cli/kanban.py` to the prior
commit (test file kept as-is) and reran the new test alone:

```
FAILED tests/tools/test_kanban_tools.py::test_request_review_goal_mode_note_survives_long_body
AssertionError: {'error': 'Goal review handoff rejected by judge: no reviewer approval evidence yet. ...'}
```

Then restored the fix and reran — passes (`34 passed`, full suite green).

Also ran:
- `python -m ruff check tools/kanban_tools.py hermes_cli/kanban.py tests/tools/test_kanban_tools.py` → `All checks passed!`
- `python scripts/check-windows-footguns.py tools/kanban_tools.py hermes_cli/kanban.py tests/tools/test_kanban_tools.py` → `✓ No Windows footguns found (3 file(s) scanned).`
- `git diff --check` → clean

### Code

- [x] I searched for existing PRs to make sure this isn't a duplicate (checked #95181, #97429, #46485, #86732 — none touch the goal-judge gate on `kanban_request_review`)

## AI assistance disclosure

This PR was prepared by an autonomous Claude-based coding agent (Claude Code) operating against this repository, including the diagnosis, fix, and test. A human reviews and controls what merges.

### DIFF
--- a/hermes_cli/kanban.py
+++ b/hermes_cli/kanban.py
@@ -2311,7 +2311,20 @@ def _worker_run_id_for(task_id: str) -> Optional[int]:
         return None
 
 
-def _goal_mode_handoff_rejection(task: Optional[kb.Task], evidence: str) -> Optional[str]:
+# Kept in sync with tools/kanban_tools.py's copy of this note (#98160).
+_REVIEW_READINESS_NOTE = (
+    "\n\nJudging note: this checks whether the implementation work described "
+    "above is finished and ready to hand off to a reviewer — it does not "
+    "check whether the card as a whole is done. If the card's own criteria "
+    "call for a reviewer to approve or close out the work, treat that as a "
+    "later step outside this check: do not withhold DONE merely because "
+    "reviewer/approval evidence is absent."
+)
+
+
+def _goal_mode_handoff_rejection(
+    task: Optional[kb.Task], evidence: str, *, handoff: str = "complete"
+) -> Optional[str]:
     """Apply the goal judge to every terminal worker handoff, including review."""
     if task is None or not task.goal_mode:
         return None
@@ -2326,11 +2339,15 @@ def _goal_mode_handoff_rejection(task: Optional[kb.Task], evidence: str) -> Opti
 
     from hermes_cli.goals import judge_goal
 
+    goal = f"{task.title}\n\n{task.body or ''}".strip()
+    if handoff == "review":
+        goal = f"{goal}{_REVIEW_READINESS_NOTE}"
+
     verdict = "done"
     reason = ""
     try:
         verdict, reason, _, _, _ = judge_goal(
-            goal=f"{task.title}\n\n{task.body or ''}".strip(),
+            goal=goal,
             last_response=evidence.strip(),
         )
     except Exception as judge_exc:
@@ -2535,6 +2552,7 @@ def _cmd_request_review(args: argparse.Namespace) -> int:
         rejection = _goal_mode_handoff_rejection(
             kb.get_task(conn, tid),
             summary or "",
+            handoff="review",
         )
         if rejection is not None:
             print(
--- a/hermes_cli/kanban.py
+++ b/hermes_cli/kanban.py
@@ -2341,6 +2341,12 @@ def _goal_mode_handoff_rejection(
 
     goal = f"{task.title}\n\n{task.body or ''}".strip()
     if handoff == "review":
+        # judge_goal truncates its goal argument to 2000 chars before sending
+        # it to the judge; reserve room here so a long title/body can't push
+        # the note itself past that limit and silently drop it (#98160).
+        budget = 2000 - len(_REVIEW_READINESS_NOTE)
+        if len(goal) > budget:
+            goal = goal[:budget]
         goal = f"{goal}{_REVIEW_READINESS_NOTE}"
 
     verdict = "done"
--- a/tests/tools/test_kanban_tools.py
+++ b/tests/tools/test_kanban_tools.py
@@ -264,6 +264,69 @@ def test_request_review_goal_mode_asks_judge_about_implementation_not_review(
     assert "Goal completion rejected by judge" in d2.get("error", "")
 
 
+def test_request_review_goal_mode_note_survives_long_body(monkeypatch, tmp_path):
+    """A goal-mode card with a long title+body (routine: task.body has an
+    8 KB budget elsewhere) must not silently lose the review-scoping note to
+    judge_goal's own 2000-char truncation. Regression for #98160 follow-up:
+    the note used to be appended AFTER title+body, so long cards consumed
+    the whole 2000-char budget before the note was ever reached.
+
+    This deliberately does NOT mock ``judge_goal`` itself (that would bypass
+    its real ``_truncate(goal, 2000)`` call and pass regardless of the bug)
+    — it mocks the LLM transport underneath so the real truncation runs and
+    the prompt actually sent to the model can be inspected.
+    """
+    from unittest.mock import patch
+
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** Breaks goal-mode same-card review deadlock (#98160) by reframing `kanban_request_review` judge question from "is whole card (incl. review) done?" to "is implementation ready to hand off?", while keeping `kanban_complete` strictly gated. Budget-aware truncation preserves the reframing note.

**What it does**
- `tools/kanban_tools.py:215` / `hermes_cli/kanban.py:11` introduces `_REVIEW_READINESS_NOTE` ("do not withhold DONE merely because reviewer/approval evidence is absent") and extends `_goal_mode_handoff_rejection(task, evidence, *, handoff="complete")` — for `handoff=="review"` reserves `2000 - len(note)` budget, truncates `goal = title+"\n\n"+body` to budget, then appends note before `judge_goal`. `_cmd_request_review` and `_handle_request_review` pass `handoff="review"`.
- Tests `tests/tools/test_kanban_tools.py:66` (`goal_mode_asks_judge_about_implementation_not_review`) uses mock judge sensitive to note presence to prove review passes but complete still fails; `note_survives_long_body` with 60-bullet long body + real `call_llm` mock proves truncation preserves note (budget math validated against `judge_goal`'s own `_truncate(goal,2000)`).

**Non-blocking notes**
- `goal` budget uses `len(note)` (Python chars) not encoded bytes — matches `judge_goal` truncation which also uses char length; consistent.
- Fix only reframes review handoff — vague/absent evidence still rejected (note says "checks whether implementation ... is finished"), so quality gate not weakened.
- `keep in sync` comment between two copies is fragile; consider extracting to shared constant if churn continues.

PR #98261, opened 2026-08-30
fix(models): resolve Ollama Cloud probe credentials through the auth-store seam

### LINKED ISSUE #98243: Desktop model picker omits models present in the Ollama Cloud cache (e.g. deepseek-v4-flash:0731)
## Bug description

The macOS desktop app's model picker does not show models that are present in the local Ollama Cloud model cache. Example: `deepseek-v4-flash:0731` is listed in `~/.hermes/ollama_cloud_models_cache.json` and is the currently selected model on the `ollama-cloud` provider, but it does not appear in the picker. The CLI picker (`hermes model`) shows the full live catalog correctly.

## Steps to reproduce

1. Configure `ollama-cloud` as the provider and select `deepseek-v4-flash:0731`.
2. Open the model picker in the desktop app.
3. Search for `deepseek-v4-flash`.
4. Expected: the model appears (it is in the cache and is the current selection).
5. Actual: the model is missing from the list.

## Evidence

- `~/.hermes/ollama_cloud_models_cache.json` contains `{"models": [23 models including "deepseek-v4-flash:0731", "mistral-large-3:675b", "glm-5.3", ...], "cached_at": <fresh timestamp>}` — correct cache shape, not stale.
- The active session runs on `ollama-cloud` / `deepseek-v4-flash:0731`, so the provider is authenticated and current.
- `hermes model` (CLI) lists the model fine — the backend catalog is not the problem.
- The desktop picker requests `/api/model/options` with explicit-only provider rows; the model does not appear in the `ollama-cloud` row even though the cache that feeds it is complete.

## Expected behavior

The desktop picker should show every model the active provider serves, matching the CLI picker. A model that is both cached and currently selected must never be absent from the picker.

## Environment

- macOS desktop app
- Provider: `ollama-cloud`

## Related

- #70628 — Model picker: provider and model are misaligned (macOS desktop app)

liuhao1024:
## What does this PR do?

Fixes the desktop model picker dropping models that are present in the Ollama Cloud cache (#98243). The root cause is that the Ollama Cloud live `/v1/models` probe only consulted `os.environ['OLLAMA_API_KEY']`. Desktop/service processes are launched from the GUI and inherit no shell exports, so a key the user logged in with (saved in `~/.hermes/.env` or the credential pool) never reached the probe there. When the cache went stale, the fetch regressed to the models.dev subset and — worse — **overwrote** the complete cache with that subset and refreshed `cached_at`, hiding cached models (including the currently selected one, e.g. `deepseek-v4-flash:0731`) from the desktop picker while the CLI picker, running from the user's shell, still listed them.

Two changes:

1. `provider_model_ids('ollama-cloud')` now resolves `api_key`/`base_url` through `resolve_api_key_provider_credentials` (registry env vars — preferring `~/.hermes/.env` over the inherited process env — then the credential pool), the same seam the `nous` and `stepfun` branches in this function already use. The bare `os.environ` fallback inside `fetch_ollama_cloud_models` is preserved as the last resort.
2. `fetch_ollama_cloud_models` no longer lets an empty live probe overwrite a previously fetched cache with the models.dev subset: when the probe comes back empty, the stale entry (TTL-ignored) seeds the merge, so a failed probe can only ever add models. A successful live probe remains fully authoritative.

## Related Issue

Fixes #98243

## Type of Change

## Changes Made

- `hermes_cli/models.py` — `provider_model_ids` ollama-cloud branch: resolve the probe credential via the auth-store seam instead of the process-env-only fallback.
- `hermes_cli/models.py` — `fetch_ollama_cloud_models`: seed the merge with the stale cache entry when the live probe is empty, so the models.dev subset can no longer silently replace a complete cached catalog.
- `tests/hermes_cli/test_ollama_cloud_provider.py` — regression tests: auth-store key reaches the probe; auth-resolution failure degrades to the env path; a failed probe preserves cached models (in the return value and on disk); a successful probe still replaces the cache.

## How to Test

1. `python -m pytest tests/hermes_cli/test_ollama_cloud_provider.py tests/hermes_cli/test_models.py tests/hermes_cli/test_cached_fetch_api_models.py tests/hermes_cli/test_configured_builtin_models.py tests/hermes_cli/test_model_cache_swr.py -q` — Observed result: 142 passed (27 + 102 + 13).
2. Simulate the desktop environment: with `OLLAMA_API_KEY` unset in the process env but a resolvable key in the auth store, `provider_model_ids('ollama-cloud')` should pass the resolved key to the live probe — covered by `test_provider_model_ids_passes_auth_store_key_to_probe`, should pass.
3. With no resolvable credential and a stale on-disk cache containing e.g. `deepseek-v4-flash:0731`, `fetch_ollama_cloud_models(force_refresh=True)` should keep that model in both its return value and the persisted cache — covered by `test_failed_probe_keeps_cached_models`, should pass.

### Code

## For New Skills

N/A — no skill changes in this PR.

## Screenshots / Logs

```
$ python -m pytest tests/hermes_cli/test_ollama_cloud_provider.py -q
27 passed in 11.39s
```

### DIFF
--- a/hermes_cli/models.py
+++ b/hermes_cli/models.py
@@ -4043,7 +4043,29 @@ def provider_model_ids(provider: Optional[str], *, force_refresh: bool = False) 
         # safe source for the chat picker, including its empty/failure result.
         return _fetch_deepinfra_models(force_refresh=force_refresh) or []
     if normalized == "ollama-cloud":
-        live = fetch_ollama_cloud_models(force_refresh=force_refresh)
+        # Resolve the credential through the auth-store seam (registry env
+        # vars, preferring ~/.hermes/.env over the inherited process env,
+        # then the credential pool) instead of leaving the probe on
+        # fetch_ollama_cloud_models()'s bare os.environ fallback: desktop and
+        # service processes are launched from the GUI and inherit no shell
+        # exports, so a key the user logged in with never reached the live
+        # probe there and every stale refresh regressed the cache to the
+        # models.dev subset (#98243).
+        ollama_cloud_key = ""
+        ollama_cloud_base = ""
+        try:
+            from hermes_cli.auth import resolve_api_key_provider_credentials
+
+            creds = resolve_api_key_provider_credentials("ollama-cloud")
+            ollama_cloud_key = str(creds.get("api_key") or "").strip()
+            ollama_cloud_base = str(creds.get("base_url") or "").strip()
+        except Exception:
+            pass
+        live = fetch_ollama_cloud_models(
+            api_key=ollama_cloud_key or None,
+            base_url=ollama_cloud_base or None,
+            force_refresh=force_refresh,
+        )
         if live:
             return live
     if normalized in ("openai", "openai-api"):
@@ -6445,6 +6467,20 @@ def fetch_ollama_cloud_models(
     if live_models or mdev_models:
         seen: set[str] = set()
         merged: list[str] = []
+        if not live_models:
+            # An empty live probe (no resolvable credential, network outage)
+            # means models.dev is the only fresh source here — and it is a
+            # SUBSET of the live catalog. Letting it overwrite a previously
+            # fetched cache would silently drop every model models.dev hasn't
+            # synced yet, and the refreshed cached_at would then hide the
+            # loss from every picker for the next TTL window (#98243). Seed
+            # the merge with the stale entry so a failed probe can only ever
+            # ADD models.
+            stale_entry = _load_ollama_cloud_cache(ignore_ttl=True)
+            for m in (stale_entry or {}).get("models") or []:
+                if m and m not in seen:
+                    seen.add(m)
+                    merged.append(m)
         for m in live_models:
             if m and m not in seen:
                 seen.add(m)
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Correctly routes Ollama Cloud live probe through auth-store seam and prevents empty-probe cache regression to the smaller models.dev subset (#98243).

**What it does**
- `hermes_cli/models.py:8` `provider_model_ids` for `ollama-cloud` now resolves `resolve_api_key_provider_credentials("ollama-cloud")` → `api_key`/`base_url`, falls back to `os.environ` on exception, and passes explicitly to `fetch_ollama_cloud_models(api_key=..., base_url=..., force_refresh=...)`.
- `models.py:40` `fetch_ollama_cloud_models` when `not live_models` but `mdev_models` present: seeds `merged` from `_load_ollama_cloud_cache(ignore_ttl=True)` (stale cache) before merging `live_models`+`mdev_models`, then persists — so failed probe can only add models, not drop cached ones; successful probe (`live_models` non-empty) still replaces stale cache (dropped models stay dropped).
- Tests `test_ollama_cloud_provider.py:69` verify credential passed from auth-store, fallback on failure, preservation of `["deepseek-v4-flash:0731","mistral-large-3:675b"]` alongside `glm-5` after empty live probe, and successful probe replacement.

**Non-blocking notes**
- `try/except Exception: pass` around `resolve_api_key...` is intentionally broad (auth store unreadable → silent fallback) — matches "survives auth resolution failure" test; could log at debug but acceptable.
- Seeding from stale cache only when `not live_models` is precisely the regression window (no credential / network outage / still on shell-env path) — if `live_models` is partial (network partial fetch) it still replaces, but `fetch_api_models` either returns complete list or `[]`, so semantics hold.
- `base_url` from auth-store now propagated — ensures custom Ollama Cloud base survives desktop/server launch without env.

liuhao1024:
Thanks for the review. The broad `except` around `resolve_api_key_provider_credentials` is intentional — the auth store being unreadable should silently fall back to the shell-env path, which the fallback test pins. Agreed that `base_url` propagation through the auth-store seam is the other half of the fix; both are covered by the new tests.

PR #98264, opened 2026-08-30
fix(desktop): show the variant tag in model picker rows

### LINKED ISSUE #98122: [Feature]: Desktop model picker rows should show the provider-prefixed canonical model slug (aggregator routes are indistinguishable)
### Problem

In the Hermes **desktop** app, the composer model picker renders every model as a prettified display name derived from `modelDisplayParts()` (`apps/desktop/src/lib/model-status-label.ts`). `modelBaseId()` slices at the **last** `/` and the menu takes only `.name` — so the canonical model slug and any provider/upstream prefix are never shown. When a provider is an aggregator that serves the same model through multiple upstream routes, the rows become literally **indistinguishable**.

**Real-world example (single custom provider `omni`, one section, no sub-grouping):**

| Rendered label | Actual model IDs in the `model.options` payload |
|---|---|
| Grok 4.6 | `gc/grok-4.6`, `grok-cli/grok-4.6` |
| Claude Sonnet 4 6 | `antigravity/claude-sonnet-4-6`, `no-think/antigravity/claude-sonnet-4-6` |
| Claude Opus 4 6 Thinking | `antigravity/claude-opus-4-6-thinking`, `no-think/antigravity/claude-opus-4-6-thinking` |
| Step 3.7 Flash Free | `kg/stepfun/step-3.7-flash:free`, `kilo-gateway/stepfun/step-3.7-flash:free` |
| Longcat 2.0 Free | `kg/meituan/longcat-2.0-free`, `kilo-gateway/meituan/longcat-2.0-free` |
| DeepSeek V4 Flash 0731 | `infrex/deepseek-v4-flash-0731`, `kilo/deepseek-v4-flash-0731` (cross-provider, same problem) |

Two rows can even render the same label **within one provider group** — there is no way to tell which route you are picking without leaving the UI.

### Root cause

1. `apps/desktop/src/lib/model-status-label.ts:74` — `modelDisplayParts()` correctly extracts variant suffixes into a `tag` field, **but** `apps/desktop/src/app/shell/model-catalog-menu.tsx:405` renders only `modelDisplayParts(family.id).name` and **discards the tag**, so distinct IDs collapse to identical row labels.
2. Provider group headers (`model-catalog-menu.tsx:386`) render only `group.provider.name`; the canonical `provider.slug` is never rendered anywhere on a row or header.
3. The data needed is **already on the wire**: gateway `model.options` returns `{providers: [{slug, name, models, aliases, ...}]}` — the picker simply doesn't display it. Search matches on `provider.slug` + `family.id` (`:554-557`) while the row itself hides them.

### Expected behavior

- Picker rows should show enough of the canonical slug to disambiguate — e.g. secondary/tertiary text with the full `model` id (and `provider.slug` context for user-defined providers where the row slug is only the bare config key).
- At minimum, the variant `tag` extracted by `modelDisplayParts()` should not be dropped in the list view (the Edit-models visibility dialog renders it correctly — the menu does not).

### Related issues (partial coverage only)

- #58566 — sub-provider drill-down for aggregators (CLI-scoped; PR #58580 open/unmerged)
- #88881 — user-configurable display labels (inverse fix: manual labels instead of showing wire data)
- #74391 / PR #74405 — Fireworks router variants (one provider's path case)
- #84069 — status pill shows generic "custom"
- #94706 — picker strips `-fast` from custom provider names (same collapse mechanism)

None of these covers "show the provider-prefixed canonical slug in desktop picker rows".

### Affected Component

Desktop app (composer model picker + model status pill)

### Environment

- Hermes desktop (apps/desktop), current master (`a08dfab`)
- Reproducible with any custom provider whose catalog contains slash-namespaced upstream IDs (OpenRouter-style, OmniRouter, llm-proxy, kilo, etc.)

chelsealong:
## What does this PR do?

Restores the variant tag (`Fast` / `Thinking` / `Preview` / `Latest`) that
`modelDisplayParts()` already extracts, but that the desktop model picker
(`model-catalog-menu.tsx`) silently drops when rendering each row. A base
model and its `-thinking` (etc.) sibling currently render **identical row
text**, so they're indistinguishable in the picker — even though the exact
same tag is already rendered correctly one dialog over, in the Edit Models
visibility list (`model-visibility-dialog.tsx`).

This is the "at minimum" fix called out in the issue: showing the full
provider-prefixed canonical slug is a larger product decision left to a
maintainer, but not dropping data the picker already computes is not.

## Related Issue

Fixes #98122

## Type of Change

## Changes Made

- `apps/desktop/src/app/shell/model-catalog-menu.tsx`: destructure `{ name, tag }`
  from `modelDisplayParts(family.id)` instead of only `.name`, and render the
  tag next to the name (same treatment `model-visibility-dialog.tsx` already
  gives it).
- `apps/desktop/src/app/shell/model-catalog-menu.test.tsx`: new test asserting
  a base model and its `-thinking` variant produce distinguishable row labels.

## How to Test

1. `cd apps/desktop && npx vitest run src/app/shell/model-catalog-menu.test.tsx`
2. Open the composer model picker for a provider whose catalog includes a
   `-thinking`/`-preview`/`-latest` sibling of a model already shown (e.g.
   `claude-opus-4.8` + `claude-opus-4.8-thinking`) — the two rows now read
   differently instead of both showing `Opus 4.8`.

### Proof

Test added in this PR, run against the code on `main` (fix reverted via
`git checkout HEAD~1 -- apps/desktop/src/app/shell/model-catalog-menu.tsx`):

```
FAIL  |ui| src/app/shell/model-catalog-menu.test.tsx > the catalog owns model curation > keeps a base model and its -thinking variant distinguishable in the row label
AssertionError: expected 1 to be 2 // Object.is equality
- Expected: 2
+ Received: 1
```

Same test with the fix applied:

```
 Test Files  1 passed (1)
      Tests  4 passed (4)
```

Also ran, with the fix applied:
- `npx eslint src/app/shell/model-catalog-menu.tsx src/app/shell/model-catalog-menu.test.tsx` — clean
- `npx tsc -p . --noEmit` — clean (whole `apps/desktop` project)
- `npx vitest run --project ui src/app/shell` — 10 files / 74 tests passed

(The full `npm run test:ui` suite — ~2500+ test files across the whole
desktop app — did not finish inside a 10-minute foreground budget in this
sandbox; that appears to be pre-existing suite runtime, not something this
2-line change affects, and the directly relevant directory finished cleanly
in ~14s.)

### Code

- [ ] I've tested on my platform — tested via jsdom/vitest only, not a live
      Electron build

- [x] N/A — no docs, config keys, or tool schemas touched

---
Drafted with AI assistance (Claude Code); test run output above is real,
from this sandbox.

### DIFF
--- a/apps/desktop/src/app/shell/model-catalog-menu.test.tsx
+++ b/apps/desktop/src/app/shell/model-catalog-menu.test.tsx
@@ -105,4 +105,18 @@ describe('the catalog owns model curation', () => {
 
     expect($modelVisibilityOpen.get()).toBe(true)
   })
+
+  it('keeps a base model and its -thinking variant distinguishable in the row label', async () => {
+    getGlobalModelOptions.mockResolvedValue({
+      providers: [{ models: ['claude-opus-4.8', 'claude-opus-4.8-thinking'], name: 'Anthropic', slug: 'anthropic' }]
+    })
+
+    renderMenu()
+
+    const rows = await screen.findAllByRole('menuitem', { name: /Opus 4\.8/i })
+    const labels = rows.map(row => row.textContent)
+
+    expect(labels).toHaveLength(2)
+    expect(new Set(labels).size).toBe(2)
+  })
 })
--- a/apps/desktop/src/app/shell/model-catalog-menu.tsx
+++ b/apps/desktop/src/app/shell/model-catalog-menu.tsx
@@ -406,7 +406,7 @@ export function ModelCatalogMenu({
                         : null
 
                     const isCurrent = activeId !== null
-                    const name = modelDisplayParts(family.id).name
+                    const { name, tag } = modelDisplayParts(family.id)
                     const caps = group.provider.capabilities?.[family.id]
 
                     // Effective settings for this row: the live choice when it's
@@ -455,6 +455,7 @@ export function ModelCatalogMenu({
                         >
                           <span className="min-w-0 flex-1 truncate">
                             <HighlightMatches query={search} text={name} />
+                            {tag ? <span className="text-(--ui-text-tertiary)"> {tag}</span> : null}
                             {meta ? <span className="text-(--ui-text-tertiary)"> {meta}</span> : null}
                           </span>
                           {isCurrent ? (


Enough1122:

**Overall:** One-line restoration of variant tag in model picker rows. `modelDisplayParts` already extracted `tag` (`Fast`/`Thinking`/`Preview`/`Latest`) but `model-catalog-menu.tsx` rendered only `name`; siblings like `-thinking` were indistinguishable.

**What it does**
- `apps/desktop/src/app/shell/model-catalog-menu.tsx:33` changes `const name = modelDisplayParts(family.id).name` → `const { name, tag } = modelDisplayParts(family.id)` and renders `{tag ? <span className="text-(--ui-text-tertiary)"> {tag}</span> : null}` alongside `meta`.
- Test `model-catalog-menu.test.tsx:10` asserts `claude-opus-4.8` vs `claude-opus-4.8-thinking` produce two `menuitem` rows with distinct `textContent` (Set size 2).

**Non-blocking notes**
- Tag rendering uses same tertiary typography as `meta` — visually consistent with visibility list (`model-visibility`) that already rendered tag correctly.
- No behavioral change to selection/effort logic.

No blocking issues.

PR #98269, opened 2026-08-30
fix(webhook): restore provider contracts and safe readiness

aviyashchin:
## Summary
- accept Circleback/Attio provider HMAC headers and provider event headers
- add side-effect-free GET readiness with profile and disabled-route authorization
- deduplicate exact body retries without suppressing changed events that share a resource ID
- accept plugin, MCP server, and `no_mcp` passthrough entries in platform-toolset migration validation

## Verification
- `pytest -q tests/gateway/test_webhook_adapter.py tests/hermes_cli/test_toolset_validation.py` (49 passed)
- `ruff check` on all changed files
- `git diff --check`
- three independent review rounds; final verdict PASS
- live Circleback provider request returned 202 and produced exactly one wake receipt

## Notes
No secrets or provider payload contents are logged. Payload fallback keys include only route, scalar ID, and SHA-256 of the exact raw body.

### DIFF
--- a/gateway/platforms/webhook.py
+++ b/gateway/platforms/webhook.py
@@ -68,6 +68,29 @@ from gateway.response_filters import is_autonomous_silence_response
 logger = logging.getLogger(__name__)
 
 
+def _payload_delivery_id(
+    route_name: str, payload: Any, body: bytes
+) -> Optional[str]:
+    """Return a retry key for one exact body carrying a stable resource ID.
+
+    A provider's top-level ``id`` is often a resource ID rather than an event
+    delivery ID. Include the raw-body digest so exact retries deduplicate while
+    distinct events about the same resource continue through the pipeline.
+    """
+    if not isinstance(payload, dict):
+        return None
+    payload_id = payload.get("id")
+    if isinstance(payload_id, bool) or not isinstance(payload_id, (str, int)):
+        return None
+    if isinstance(payload_id, str) and not payload_id:
+        return None
+    return "payload-id:" + json.dumps(
+        [route_name, payload_id, hashlib.sha256(body).hexdigest()],
+        separators=(",", ":"),
+        ensure_ascii=True,
+    )
+
+
 def _is_webhook_silence_response(content: Any) -> bool:
     """Whether an agent response means "deliberately say nothing".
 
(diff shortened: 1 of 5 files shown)


Enough1122:

**Overall:** Solid webhook hardening across provider contracts, readiness, dedup, and config validation. Each change is narrowly scoped with matching tests and reuses existing primitives.

**What it does**
- `gateway/platforms/webhook.py:9` adds `_payload_delivery_id(route,payload,body)` → `payload-id:<json[route, id, sha256(body)]>` so exact retries of a provider like Circleback (stable payload `id`) deduplicate while distinct events sharing resource ID (`meeting.created` vs `meeting.completed` with same `id`) have different body digest and both deliver.
- Extends `_validate_signature` generic branch to `Attio-Signature`/`X-Attio-Signature`/`X-Signature` plus `sha256=` prefix stripping; warns `X-Webhook-Signature` legacy only once per route.
- Adds event-type fallthrough `Linear-Event` then `X-Event-Type`; adds `GET /webhooks/{route}` + `GET /p/{profile}/webhooks/{route}` ` _handle_webhook_readiness` — side-effect-free, checks `_reload_dynamic_routes`, `resolve_request_profile`, `get_route`, `route_allows_profile`, `enabled` → `200 {"ready"}` / `404` / `403`; POST handler now prioritizes trusted `X-GitHub-Delivery`/`svix-id`/`X-Request-ID` then falls back to `_payload_delivery_id` then timestamp.
- `hermes_cli/config.py:168` / `toolset_validation.py:12` extends `validate_platform_toolsets` with `additional_valid_names` (`no_mcp` + `known_plugin_toolsets` + `mcp_servers` keys from raw config) so plugin/MCP passthrough entries no longer trigger false warnings during `migrate_config`.

**Non-blocking notes**
- Delivery-ID chain now empty `X-Request-ID` falls through to payload-id — correct; prior `or str(int(time...))` default is now last resort only, reducing spurious dedup collisions.
- Readiness tests mirror `connect()` router setup exactly (`_create_app` helper) — good; ensure docs advertise `GET` as probe not requiring signature.
- `toolset_validation` addition is additive/unbreaking — static `validate_toolset` still authoritative, dynamic set is union.

Issue #97983, opened 2026-08-29
fix: Linux desktop entry Exec= broken on uv-managed venvs (symlinked venv python)


mojtabazn:
## Summary

After `hermes update` on Linux, the Hermes Desktop icon in the app menu fails to launch silently (`Terminal=false`). The generated `~/.local/share/applications/hermes.desktop` contains a broken `Exec=` line that prefixes the **outer uv base interpreter**, which has no `hermes_cli` in its site-packages:

```
Exec=/home/<user>/.local/share/uv/python/cpython-3.13.15-linux-x86_64-gnu/bin/python3.13 /home/<user>/.hermes/hermes-agent/venv/bin/hermes desktop
```

Running that command manually fails with:

```
Traceback (most recent call last):
  File ".../venv/bin/hermes", line 4, in <module>
    from hermes_cli.main import main
ModuleNotFoundError: No module named 'hermes_cli'
```

## Root cause

`hermes_cli/linux_desktop_entry.py`, `_needs_interpreter()` (added in #92318 / aff5125f8), decides whether the resolved `hermes` console script needs an interpreter prefix by checking whether the script's shebang contains the **resolved** directory of the running interpreter:

```python
exe_dir = str(Path(sys.executable).resolve().parent)
return exe_dir not in shebang
```

uv-managed venvs symlink `venv/bin/python` into the shared base interpreter (`~/.local/share/uv/python/<version>/bin/python3.13`). So `sys.executable` *resolves* to the uv directory, while the console script's shebang names the *venv* directory (`#!/path/to/venv/bin/python3`). The substring check fails → Hermes prefixes the outer uv python → the launcher dies before Electron starts.

This also explains why it regressed "after an update": the update recreated the venv under uv with the symlinked layout. On venvs where `venv/bin/python` is a real copy (not a symlink), the naive path check works, which is why this isn't widespread yet.

## Environment

- Hermes v0.20.6 (git install), Linux (KDE Plasma / Wayland)
- venv created by uv 0.12.1: `venv/bin/python` is a symlink into `~/.local/share/uv/python/cpython-3.13.15-linux-x86_64-gnu/bin/python3.13`
- `venv/bin/hermes` shebang: `#!/path/to/hermes-agent/venv/bin/python3`

## Proposed fix

Resolve the shebang's interpreter and compare its real directory with the running interpreter's real directory. If they match, the script already runs in this environment and no prefix is needed:

```python
[...]

Issue #94564, opened 2026-08-25
Linux desktop entry broken in uv venvs: Exec resolves sys.executable through the venv symlink to the base interpreter


deepanaishtaweera:
## Summary

On Linux, the `hermes.desktop` launcher entry written by `hermes desktop` silently fails to open the app when Hermes runs in a **uv-created venv**. The generated `Exec=` line points at the *base* uv interpreter instead of the venv interpreter, so the launch dies on the first third-party import (`ModuleNotFoundError: No module named 'yaml'`) — invisibly, since the entry has `Terminal=false`.

## Root cause

`hermes_cli/linux_desktop_entry.py` builds the `Exec=` line from `sys.executable`, but calls `.resolve()` on it:

```python
argv = [str(Path(sys.executable).resolve()), str(resolved), "desktop"]   # line 81
...
argv = [str(Path(sys.executable).resolve()), "-m", "hermes_cli.main", "desktop"]  # line 85
...
exe_dir = str(Path(sys.executable).resolve().parent)  # line 109, _needs_interpreter
```

In a uv venv, `venv/bin/python` is a **symlink** to the managed base interpreter:

```
venv/bin/python -> ~/.local/share/uv/python/cpython-3.11-linux-x86_64-gnu/bin/python3.11
```

`.resolve()` follows that symlink right out of the venv. The written entry becomes:

```
Exec=/home/<user>/.local/share/uv/python/cpython-3.11.16-linux-x86_64-gnu/bin/python3.11 /home/<user>/.hermes/hermes-agent/hermes desktop
```

That base interpreter has no venv site-packages, so the launch crashes immediately:

```
Traceback (most recent call last):
  File "/home/<user>/.hermes/hermes-agent/hermes", line 10, in <module>
    from hermes_cli.main import main
  File ".../hermes_cli/main.py", line 715, in <module>
    from hermes_cli.config import get_hermes_home
  File ".../hermes_cli/config.py", line 331, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'
```

The comment above line 81 even says the intent is "sys.executable is the interpreter actually running Hermes (the venv one)" — the `.resolve()` defeats exactly that. This is the same failure mode the `_needs_interpreter` shim was added to prevent (referenced there as #90292), reintroduced through the symlink.

[...]

Issue #92882, opened 2026-08-23
[Bug]: `hermes desktop` fails with ModuleNotFoundError (yaml/dotenv) when launched with a uv-managed CPython outside the install venv


mylittlebotclaw:
### Bug Description

Running `hermes desktop` via a **uv-managed interpreter that is not the install venv** (`~/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/bin/python3.11 ~/.hermes/hermes-agent/hermes desktop`) crashes at import time with `ModuleNotFoundError`, because that interpreter has none of Hermes' dependencies (`yaml`, `dotenv`, ...). The same happens through the generated `.desktop` entry, whose `Exec=` resolves to this bare uv interpreter instead of the install venv's Python.

Related prior reports / PRs on the same symptom family:
- #90292 — Linux desktop entry generated with non-runnable Exec
- #91504 — Desktop entry Exec= points at raw repo script; env python3 resolves outside venv (closed as duplicate)
- #80439 — auto-generated hermes.desktop uses wrong Exec path, breaking KDE taskbar pinning
- #92090 (open PR) — keep the venv interpreter in Exec= instead of its symlink target
- #77638 (open PR) — probe-gate and version-order the POSIX Python fallback

### Steps to Reproduce

1. Have a uv-managed CPython (e.g. installed via `uv python install`), with no Hermes deps installed in it
2. Run: `~/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/bin/python3.11 ~/.hermes/hermes-agent/hermes desktop`
3. Also reproducible via the generated launcher `~/.local/share/applications/hermes.desktop`, whose `Exec=` points at the same interpreter/script pair

### Expected Behavior

The launcher (and any invocation of the repo's `hermes` script) should run under the install venv interpreter that actually has `hermes_cli`'s dependencies installed, or fail with a clear actionable error — never a bare `ModuleNotFoundError: No module named 'yaml'`.

### Actual Behavior

Three consecutive tracebacks, all import-time failures:

```
Traceback (most recent call last):
  File "~/.hermes/hermes-agent/hermes", line 10, in <module>
    from hermes_cli.main import main
  File "~/.hermes/hermes-agent/hermes_cli/main.py", line 723, in <module>
    from hermes_cli.config import get_hermes_home
  File "~/.hermes/hermes-agent/hermes_cli/config.py", line 331, in <module>
    import yaml
[...]

PR #98279, opened 2026-08-30
fix(linux): keep venv python in desktop entry Exec

fischerfp:
## Problem

On uv-managed installs the generated Linux desktop entry is unusable: clicking the icon does nothing at all. Launching `hermes desktop` from a terminal works fine, which makes the failure look like a desktop-environment issue rather than a Hermes one.

It fails silently because the entry sets `Terminal=false`, so the traceback never reaches the user.

## Cause

`resolve_exec_command()` builds the `Exec=` line with `Path(sys.executable).resolve()`.

On a uv-managed install, `venv/bin/python` is a **symlink** to the base interpreter. `.resolve()` follows that symlink, so the `Exec=` line ends up pointing at the bare interpreter outside the venv — undoing exactly the venv membership the function exists to preserve. The comment above the line states the intent correctly ("sys.executable is the interpreter actually running Hermes (the venv one), so prefix it explicitly"); `.resolve()` defeats it.

The launcher then dies on the first third-party import:

```
ModuleNotFoundError: No module named 'yaml'
```

## Reproduction

Install via the uv-based installer, then:

```console
$ venv/bin/python -c "import sys; from pathlib import Path; \
    print(sys.executable); print(Path(sys.executable).resolve())"
~/.hermes/hermes-agent/venv/bin/python
~/.local/share/uv/python/cpython-3.11.16-linux-x86_64-gnu/bin/python3.11

$ venv/bin/python -c "import yaml"                      # ok
$ ~/.local/share/uv/python/.../bin/python3.11 -c "import yaml"
ModuleNotFoundError: No module named 'yaml'
```

The second path is what lands in `Exec=`.

Running the generated `Exec=` line in a launcher-like environment reproduces the silent failure:

```console
$ env -i HOME=$HOME DISPLAY=:0 PATH=/usr/local/bin:/usr/bin:/bin \
    <resolved-interpreter> ~/.hermes/hermes-agent/hermes desktop
ModuleNotFoundError: No module named 'yaml'
```

Because the entry is regenerated on each launch, editing `~/.local/share/applications/hermes.desktop` by hand is overwritten on the next start — which is why this reads to users as "it worked once, then stopped".

## Fix

Drop `.resolve()` in the two `argv` constructions so the venv path is kept as-is. `_needs_interpreter()` is left untouched: its `.resolve()` is used for shebang comparison, which is a different question.

## Verification

With the patch, the regenerated entry reads:

```
Exec=~/.hermes/hermes-agent/venv/bin/python -m hermes_cli.main desktop
```

and `gtk-launch hermes` — the same path the icon takes — starts the app normally. Tested on Linux Mint (Cinnamon, X11), Hermes Agent v0.20.6, uv-based git install.

## Notes

Refs #90292, #94058. Also reported as #97983, #94564, and #92882 (the `ModuleNotFoundError` symptom). Confirmations on the linked issues span CachyOS/KDE, Ubuntu/GNOME and Mint/Cinnamon, so this does not look distribution-specific.

### DIFF
--- a/hermes_cli/linux_desktop_entry.py
+++ b/hermes_cli/linux_desktop_entry.py
@@ -78,11 +78,11 @@ def resolve_exec_command() -> str:
             # third-party import (#90292) — silently, since Terminal=false.
             # sys.executable is the interpreter actually running Hermes (the
             # venv one), so prefix it explicitly.
-            argv = [str(Path(sys.executable).resolve()), str(resolved), "desktop"]
+            argv = [str(sys.executable), str(resolved), "desktop"]
         else:
             argv = [str(resolved), "desktop"]
     else:
-        argv = [str(Path(sys.executable).resolve()), "-m", "hermes_cli.main", "desktop"]
+        argv = [str(sys.executable), "-m", "hermes_cli.main", "desktop"]
     return " ".join(_quote_exec_arg(a) for a in argv)
 
 


alt-glitch:

Duplicate of #95256 — both preserve the unresolved venv interpreter in the Linux desktop `Exec=` line so a uv-managed symlink does not escape Hermes site-packages. #95256 also includes regression coverage.

PR #98281, opened 2026-08-30
feat(agent): opt-in cross-tool result cache with TTL and budget caps

joshuammoss-crypto:
## Summary

Adds `agent/tool_cache.py`, an opt-in cross-tool result cache. Tools can wrap expensive read-only calls with the `@cached()` decorator to skip recomputation when the same arguments reappear within a TTL.

## What's in this PR

- `agent/tool_cache.py` — module with `@cached()`, `cache_stats()`, `clear_cache()`, `reset_stats()`. Storage: `<HERMES_TOOL_CACHE_DIR>/<tool>/<key[:2]>/<key>.json`. Atomic writes (temp + rename). TTL honoured at read time. Per-tool TTL via `cached(ttl_seconds=...)`. Custom key functions via `cached(key_from=lambda ...)`. Module-wide budget cap (entries + bytes) with LRU eviction by mtime.
- `tests/agent/test_tool_cache.py` — 10 tests covering key stability, TTL expiry, exception non-caching, unserialisable-result skipping, scoped cache clear, atomic write safety, budget enforcement, and noisy-arg dropping via `key_from`.

## Why opt-in / decorator-only

The existing `tool_executor.py` runs concurrent tool calls through hot paths (`execute_tool_calls_concurrent`, `execute_tool_calls_sequential`). Wiring the cache into those call sites needs careful handling of mutating tools, partial-failure semantics, and the `_stats` event broadcast. That's a follow-up PR.

This PR ships the library and decorator with full tests so any tool author can adopt it without touching core. It also gives reviewers a small, easy-to-audit surface to evaluate the design before the executor integration lands.

## Design decisions

- **JSON on disk, not sqlite**: smaller dep, easier to inspect, atomic write is trivial. For higher-volume caches the executor-integration PR can swap the storage layer behind the same API.
- **Module-global `_stats`**: kept simple; tests reset via `reset_stats()`.
- **`key_from` for noisy args**: drop timestamps, request IDs, trace tokens without changing the cached function signature.
- **TTL honoured at read time**: expired entries are deleted lazily; no background sweeper needed.
- **Exceptions never cached**: the wrapper's only job on exception is to re-raise.
- **Unserialisable results skipped silently**: returning the value to the caller without storing — never block a real call on a cache decision.

## Risk

Low. Purely additive. No changes to existing tool call paths. The module is imported only when `@cached` is applied.

## Searched for duplicates

- `agent/prompt_caching.py` — Anthropic prompt cache markers, unrelated concept.
- No existing `tool_cache`, `result_cache`, or similar decorator in the tree.

## Tests

```
$ pytest tests/agent/test_tool_cache.py -v
======================== 10 passed in 0.42s ========================
```

### DIFF
--- a/agent/tool_cache.py
+++ b/agent/tool_cache.py
@@ -1,0 +1,258 @@
+"""Cross-tool result cache with TTL and budget caps.
+
+Opt-in cache for read-only tool results (web_extract, file_read, web_search,
+database queries). Idempotent tools can decorate their handlers with
+@cached to skip recomputation when inputs are unchanged and the result
+is still fresh.
+
+Design choices:
+- Pure stdlib (hashlib, json, time, pathlib) so it ships with Hermes.
+- Content-addressed key: sha256(canonical_json(args)).
+- TTL is per-tool-set; default 5 minutes; overridden by tool metadata.
+- Cache entries are pruned on every write to keep disk bounded.
+- Failures are NEVER cached. Only successful returns with serializable
+  payloads are eligible.
+
+Usage::
+
+    from agent.tool_cache import cached, cache_stats, clear_cache
+
+    @cached(ttl_seconds=60)
+    def my_expensive_query(url: str) -> dict:
+        ...
+
+    # Outside-inspect:
+    cache_stats()        # {"entries": 12, "size_bytes": 8421, "hits": 4, "misses": 18}
+    clear_cache("tool:my_expensive_query")  # scoped
+    clear_cache()                            # whole cache
+"""
+
+from __future__ import annotations
+
+import functools
+import hashlib
+import json
+import os
+import sys
+import tempfile
+import time
+from collections import OrderedDict
+from pathlib import Path
+from typing import Any, Callable, Optional
+
+
+_DEFAULT_TTL = 300  # 5 minutes
+_MAX_ENTRIES = 1024
+_MAX_BYTES = 64 * 1024 * 1024  # 64 MiB
+
+
+def _cache_dir() -> Path:
+    """Return the cache root directory, creating it if missing.
+
+    Honours ``HERMES_TOOL_CACHE_DIR`` for tests and admin overrides.
+    """
+    override = os.environ.get("HERMES_TOOL_CACHE_DIR")
+    if override:
+        root = Path(override)
+    else:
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Well-designed opt-in cross-tool cache (currently unused — no call sites yet, pure additive module). Stdlib-only, content-addressed, atomic writes, TTL, and bounded budget with graceful degradation. Good observability via `cache_stats`.

**What it does**
- Adds `agent/tool_cache.py` (`cached`, `cache_stats`, `clear_cache`, `reset_stats`): key `sha256(canonical_json(args))` (sort_keys), shard `tool/<key[:2]>/<key>.json`, envelope `{value, expires_at, stored_at}`, `wait_for` TTL at read (expired → unlink), `tmp+rename` atomic write, `_enforce_budget` (oldest mtime first, caps `1024 entries` / `64 MiB`), `json.dumps(value)` serializability gate (skip sets/custom), `key_from` for noisy args.
- Tests `tests/agent/test_tool_cache.py` — 8 cases: kwargs order independence, hits/misses/stores, TTL expiry, exception not cached, unserializable skip, scoped `clear_cache`, `key_from` dropping timestamp, atomic-write temp cleanup, budget cap (mocked `1024→3`).

**Non-blocking notes**
- Module is opt-in `@cached` decorator with no auto-wrapping of tools — zero risk to existing tools until explicitly adopted; when adopted, confirm read-only/idempotent semantics (never cache failures or side-effecting tools).
- `default=str` used only for key hashing, NOT for value serialization gate — value gate uses plain `json.dumps(value)` (no `default=str`) so sets correctly bypass cache; key fallback `default=str` may coerce exotic arg types to string keys — acceptable but document that callers with non-JSON args should supply `key_from`.
- `_cache_dir` respects `HERMES_TOOL_CACHE_DIR` (tests) and `LOCALAPPDATA` fallback on Windows; `rglob` on every `cache_stats`/`_enforce_budget` is bounded by `64MiB/1024` caps but could be frequent — fine for opt-in low-throughput tools.
- `stored_at`/`expires_at` use `int(time.time())` — second granularity; `ttl_seconds=1` test manipulates `expires_at` directly, so fine.

Issue #95706, opened 2026-08-26
cronjob create with no_agent=true + script fails: TypeError '<=' not supported between str and int


falyhery:
## Bug

`cronjob(action="create", ..., no_agent=true, script=...)` fails immediately with:

```
TypeError: '<=' not supported between instances of 'str' and 'int'
```

The job is NOT created.

### Environment
- Hermes Agent v0.20.5 (2026.8.19), upstream 9aa7530f
- macOS host, profile `alpha-test`

## Steps to reproduce
1. `cronjob(action="create", name="...", schedule="0 11 * * 1", deliver="bot-chat", no_agent=true, repeat="forever", script="<any value>")`
2. Returns the TypeError immediately.

## Observed
- Error occurs for ANY `script` value: absolute path under `~/.hermes/scripts/`, filename-only, or inline bash.
- Without `script` (LLM mode), the job creates fine - but then fails at runtime because the default model free period has ended (`HTTP 404: This model free period has ended`).

## What works
- `no_agent=true` jobs are rejected without `script` (script required for no_agent).
- Manual `bash x-digest-cron.sh` works perfectly (551 bookmarks, digests regenerated).
- A launchd (macOS) LaunchAgent running the same script works as a weekly automation.

## Impact
Cannot schedule a model-free (`no_agent`) cron job that runs a script via the Hermes cron scheduler. Workaround is an OS-level scheduler (launchd), which loses the `bot-chat` delivery.

## Likely cause
A path/integer comparison in the `cronjob` tool validation when `script` is set (comparing a string field against an int threshold).

teknium1:
Fixed on main — both halves of this class:

- Create path: PR #98275 (commit 89b38ed734) — string repeat forms (`forever`/`once`/numeric) coerce at `create_job`, so every entry point (tool, CLI, API, dashboard, no_agent+script) inherits it.
- Update path: PR #98290 (commit 52e5e7c034) — `update_job` and the cronjob tool's update handler route through the same `normalize_repeat_value()` chokepoint; previously updates either raised the same TypeError or stored the raw string and broke the next `mark_job_run`.

First-submitter credit for this fix goes to @ARegalado1 (PR #7216, April), with the merged implementation authored by @andrexibiza (#78928/#77366).

PR #77366, opened 2026-08-03
fix(cron): coerce string repeat values ('forever'/'once') on create and update

andrexibiza:
Related #7142 #7216 #7424 #64520 #64724 #66824 #71987 #71993
## What / Why

`cronjob(action='create', repeat='forever')` (and the same on `update`) crashes with:

```
TypeError: '<=' not supported between instances of 'str' and 'int'
```

Both normalization sites compared the raw `repeat` value against `0` before any coercion:
- `cron/jobs.py:1315` (create_job) — `if repeat is not None and repeat <= 0:`
- `tools/cronjob_tools.py:1003` (update path) — `normalized_repeat = None if repeat <= 0 else repeat`

Models commonly pass `"forever"`, `"once"`, or `"3"` (string) because the schema advertised only an integer while the docs told users `'forever'` was valid. This is the same failure family as #7142 (`repeat='once'`), #64520 (`repeat='forever'`), #66824, #71987, #71993.

### Fix

Add `normalize_repeat_value()` in `cron/jobs.py` and use it at **both** call sites:

| input | result |
|---|---|
| `None` | `None` (infinite) |
| `"forever"` / `"infinite"` (case-insensitive) | `None` (infinite; one-shot auto-repeat turns `None` → `1`) |
| `"once"` | `None` → one-shot schedules store `times=1` |
| `"3"` / any numeric string | `int` coerced |
| `"banana"` / garbage | `None` (treated as unset, no crash) |
| `0`, negative | `None` (infinite) |
| positive int | passthrough |

The model-facing `CRONJOB_SCHEMA` `repeat` field is widened from `{"type": "integer"}` to `anyOf [integer, string enum ["forever","infinite","once"]]` so the tool contract advertises exactly what the implementation accepts (the review gap flagged on #64724). Type signatures on `create_job` and `cronjob` widen to `Optional[Union[int, str]]`.

## Repro (before)

```python
cronjob(action="create", name="test", prompt="hi", schedule="0 9 * * *", repeat="forever", deliver="origin")
# TypeError: '<=' not supported between instances of 'str' and 'int'
```

## How to test

```
bash scripts/run_tests.sh tests/cron/test_jobs.py tests/tools/test_cronjob_tools.py
```

Regression tests added:
- `tests/cron/test_jobs.py` — create_job with `"forever"`, `"infinite"`, `"once"`, `"3"`, `"banana"`, `0`, `-3`, `5`
[...]

Issue #71987, opened 2026-07-26
Bug: cronjob create/update throws TypeError '<=' not supported between instances of 'str' and 'int'


dezcalimese:
# Bug Report: `cronjob` create/update throws TypeError

## Summary
The `cronjob` tool throws a Python `TypeError` on every `create` and `update` call. The `list` action works fine, but any attempt to create or modify a cron job fails.

## Error
```
Error: '<=' not supported between instances of 'str' and 'int'
success: false
```

## Steps to Reproduce
```bash
hermes cronjob create \
  --name "Test Job" \
  --schedule "0 10 * * *" \
  --repeat forever \
  --deliver telegram \
  --prompt "Read ~/.hermes/telegram-tools/reports/telegram_digest.txt and deliver its contents" \
  --workdir ~/.hermes/telegram-tools
```

## Observed Behavior
Every call to `cronjob action=create` or `cronjob action=update` returns:
```
{"error": "'<=' not supported between instances of 'str' and 'int'", "success": false}
```

## Environment
- Hermes Agent desktop application
- macOS
- Model: Ornith-1.0-35B-MLX-oQ8 via custom provider at 127.0.0.1:8001

## Root Cause Analysis
The error is a Python type comparison issue. The `cronjob` backend code is comparing a string value (likely the `repeat` field passed as `"forever"`) to an integer using the `<=` operator, which is not supported in Python.

Possible locations:
- The `repeat` field parser/validator may not convert "forever" to a valid integer representation before comparison
- The scheduling logic may assume `repeat` is always an integer
- A callback or comparison function expects numeric types for time-related fields

## Impact
Automated cron job scheduling is completely broken. Manual execution of scripts works fine, but no new jobs can be created or modified.

## Suggested Fix
Check where the `repeat` field is compared in the cronjob backend and ensure type consistency. Convert string representations to integers before comparison, or handle "forever" as a special case.

## Additional Notes
- `cronjob action=list` works without errors
- The same error occurs regardless of the prompt, schedule, or workdir values
- The error is consistent and always the same string comparison issue

alt-glitch:
> *This was generated by AI during triage.*

Duplicate of #7142: `repeat=forever` reaches the same uncoerced numeric comparison in cronjob create/update.

andrexibiza:
Duplicate-consolidation (2-wave blind swarm, 2026-08-03).

**Canonical:** #66824 — this issue is a duplicate and should be closed in favor of #66824.

[...]

Issue #66824, opened 2026-07-18
[Bug]: cronjob create with repeat='forever' fails: '<=' not supported between instances of 'str' and 'int'


sahilthakur456111-stack:
## Summary
`cronjob create` with `repeat='forever'` (or omitting `repeat` for a recurring schedule) fails with:
`TypeError: '<=' not supported between instances of 'str' and 'int'`

## Repro
```python
cronjob(action='create', name='test', prompt='hi', schedule='0 9 * * *', repeat='forever', deliver='origin')
# -> error
```

## Workaround
Pass an integer `repeat` (e.g. `repeat=7`) — that works. The string `'forever'` (and likely the implicit default) trips a comparison against an int somewhere in the scheduler code.

## Environment
- Hermes Agent active profile: bt-agent
- Created 2026-07-18
- Discovered while setting up a daily digest cron

## Expected
`repeat='forever'` should be accepted (or the parameter should be renamed/typed more clearly). Currently the API silently accepts the string and crashes.

alt-glitch:
> *This was generated by AI during triage.*

Duplicate of #7142: `repeat='forever'` follows the same uncoerced string-versus-integer comparison path. The open canonical tracks this repeat-string failure family.

PRATHAMESH75:
Confirmed duplicate of #7142. The exact `repeat='forever'` case is already fixed by the open PR #64724, which guards the `repeat <= 0` comparison behind an `isinstance(repeat, str)` branch that maps `forever`/`infinite` → infinite (None) in `cron/jobs.py`. Posted a review there. No separate fix needed here — tracking under #7142 / #64724. Workaround in the meantime: pass an integer `repeat`.

andrexibiza:
Duplicate of #7142 (canonical for the cron repeat string-vs-int coercion family). Fixed for the whole family by PR #77366 — 'forever'/'infinite'/'once' and numeric strings now coerce correctly on both create and update paths. Superseded.

andrexibiza:
Duplicate-consolidation (2026-08-03).

This issue is part of the cron repeat-string cluster (same root cause: `repeat='forever'` hits a str-vs-int comparison in the scheduler). **Fixed and shipped** by PR #77366 — `fix(cron): coerce string repeat values ('forever'/'once') on create and update` (author: andrexibiza), MERGEABLE with all CI checks passing. It normalizes `forever`/`infinite`/`once` and numeric strings on both create and update, with 71 lines of tests.

[...]

PR #98290, opened 2026-08-30
fix(cron): repeat='forever' no longer breaks job updates (completes the string-repeat class)

teknium1:
## Summary
`repeat="forever"`/`"once"`/`"3"` now work on the UPDATE path too — the merged create-path coercion (#98275) left `update_job` storing raw strings (breaking the next `mark_job_run` with `'str' object has no attribute 'get'`) and the cronjob tool's update handler raising the original TypeError.

Completes the string-repeat class: #66824, #64520, #7142, #71987, #95706, and the update half of #77366.

## Changes
- `cron/jobs.py`: extract `normalize_repeat_value()` as the single chokepoint (shape from #77366 by @andrexibiza, keeping the garbage-rejection semantics that merged in #98275); `create_job` and `update_job` (bare-value and dict shapes, `completed` counter preserved) both route through it
- `tools/cronjob_tools.py`: update handler's `repeat <= 0` compare replaced with the shared chokepoint
- `tests/cron/test_jobs.py`: update-path regression test (sabotage-verified — fails without the fix)

## Validation
| Path | Before | After |
|---|---|---|
| `update_job({"repeat": "forever"})` | raw string stored → next `mark_job_run` AttributeError | `times: None`, counter preserved, `mark_job_run` OK |
| tool `update` with `repeat="once"` | TypeError | `times: 1` |
| `update_job({"repeat": "banana"})` | raw string stored | ValueError |
| create path | fixed in #98275 | unchanged |

Live repro: real `cron.jobs` + `tools.cronjob_tools` imports against a temp HERMES_HOME, before/after. Sabotage run: removing the update-path normalization makes the new test fail. Targeted: 183/183 passed.

## Infographic

Repeat coercion class fix infographic

### DIFF
--- a/cron/jobs.py
+++ b/cron/jobs.py
@@ -774,6 +774,36 @@ def ensure_dirs():
 # Schedule Parsing
 # =============================================================================
 
+def normalize_repeat_value(repeat: Any) -> Optional[int]:
+    """Coerce a repeat value from any entry point into ``Optional[int]``.
+
+    The tool schema exposes ``repeat`` as an integer, but agents and users
+    legitimately pass the user-facing strings ``'forever'``/``'once'`` or
+    numeric strings (``'3'``). Uncoerced strings previously died with
+    ``'<=' not supported between instances of 'str' and 'int'`` at create
+    (#66824/#64520/#7142/#71987/#95706) and were stored raw by update paths,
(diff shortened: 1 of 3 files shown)


alt-glitch:

Related: #7216 and #77366 address the same repeat-normalization family. This core-team PR completes the update path with explicit once/forever semantics and counter preservation.

PR #98291, opened 2026-08-30
fix(slack): retry as top-level message when thread root cannot accept replies

### LINKED ISSUE #98285: fix(slack): fall back when a thread root cannot accept replies
## Bug

The Slack adapter sends `thread_ts` whenever it resolves a reply target. Slack can reject that target with `cannot_reply_to_message` when the root is a non-threadable message type (for example, a Slack system event such as a channel rename). The adapter retries a Block Kit rejection without blocks, but does not handle this Slack-specific thread rejection.

## Observed behavior

`chat.postMessage` fails and the response is not delivered:

```text
SlackApiError: The request to the Slack API failed.
The server responded with: {'ok': False, 'error': 'cannot_reply_to_message'}
```

This occurred while a gateway restart auto-resumed an interrupted Slack session. The same failure also appeared after Slack channel-system messages were received.

## Expected behavior

Hermes should not lose a response because the requested thread root is non-threadable. It should either:

1. ignore Slack system/non-threadable inbound events before scheduling a response; and/or
2. on `cannot_reply_to_message`, retry the same payload as a top-level channel message (without `thread_ts`), then clean up the Assistant/thread status.

## Implementation pointer

`plugins/platforms/slack/adapter.py`, `send()` around the `chat_postMessage(**kwargs)` call (currently around lines 3028–3059).

## Environment

- Hermes Agent: 0.20.6
- macOS 15.7.4
- Slack gateway managed by launchd

liuhao1024:
## What does this PR do?

When the resolved reply target is a non-threadable Slack message type (e.g. a Slack system event such as a channel rename), `chat.postMessage` fails with `cannot_reply_to_message` and the adapter dropped the response entirely. The `send()` path already had a fallback that retried Block Kit rejections without `blocks`; this adds the symmetric fallback for thread rejections: on `cannot_reply_to_message`, retry the same payload without `thread_ts`/`reply_broadcast` as a top-level channel message. Other thread errors (e.g. `thread_not_found`) keep failing as before.

## Related Issue

Fixes #98285

## Type of Change

## Changes Made

- `plugins/platforms/slack/adapter.py`: new `_is_thread_reply_rejection()` helper detecting `cannot_reply_to_message` (mirrors the existing `_is_block_payload_rejection()`), and a fallback branch in `send()` that retries the post as a top-level message without `thread_ts`/`reply_broadcast`.
- `tests/gateway/test_slack_block_kit_adapter.py`: new `TestThreadReplyFallback` covering the fallback retry, non-recoverable thread errors still failing, and no retry when no thread target was used.

## How to Test

1. `python -m pytest tests/gateway/test_slack_block_kit_adapter.py -q` — Observed result: `14 passed` (11 pre-existing + 3 new).
2. `python -m pytest tests/gateway/ -q -k slack` — Observed result: `641 passed, 1 failed, 1 skipped`; the single failure (`test_slack.py::TestSendMultipleImagesSSRFGuards::test_batch_download_blocks_connect_time_rebind`) reproduces on a clean `upstream/main` checkout without this change.
3. New test `test_cannot_reply_retries_as_top_level_message` simulates `cannot_reply_to_message` on the first post and asserts the retry carries no `thread_ts`/`reply_broadcast` while keeping the same text — should pass.

### Code

## Screenshots / Logs

```text
tests/gateway/test_slack_block_kit_adapter.py::TestThreadReplyFallback::test_cannot_reply_retries_as_top_level_message PASSED
tests/gateway/test_slack_block_kit_adapter.py::TestThreadReplyFallback::test_other_thread_errors_still_fail PASSED
tests/gateway/test_slack_block_kit_adapter.py::TestThreadReplyFallback::test_cannot_reply_without_thread_target_does_not_retry PASSED
```

### DIFF
--- a/plugins/platforms/slack/adapter.py
+++ b/plugins/platforms/slack/adapter.py
@@ -3055,6 +3055,23 @@ class SlackAdapter(BasePlatformAdapter):
                         last_result = await self._get_client(
                             chat_id, team_id=team_id
                         ).chat_postMessage(**retry_kwargs)
+                    elif (
+                        kwargs.get("thread_ts")
+                        and self._is_thread_reply_rejection(e)
+                    ):
+                        # Non-threadable root (e.g. a Slack system event):
+                        # retry the same payload as a top-level channel
+                        # message instead of dropping the response (#98285).
+                        retry_kwargs = dict(kwargs)
+                        retry_kwargs.pop("thread_ts", None)
+                        retry_kwargs.pop("reply_broadcast", None)
+                        logger.info(
+                            "[Slack] Thread root cannot accept replies; retrying send as top-level message: %s",
+                            e,
+                        )
+                        last_result = await self._get_client(
+                            chat_id, team_id=team_id
+                        ).chat_postMessage(**retry_kwargs)
                     else:
                         raise
 
@@ -4142,6 +4159,29 @@ class SlackAdapter(BasePlatformAdapter):
             "msg_too_long",
             "too_many_blocks",
         }
+        response = getattr(error, "response", None)
+        response_get = getattr(response, "get", None)
+        if callable(response_get):
+            try:
+                if response_get("error") in recoverable_codes:
+                    return True
+            except Exception:
+                pass
+        message = str(error)
+        return any(code in message for code in recoverable_codes)
+
+    @staticmethod
+    def _is_thread_reply_rejection(error: BaseException) -> bool:
+        """Return True for Slack errors recoverable by dropping ``thread_ts``.
+
+        Some thread roots cannot accept replies — non-threadable message
+        types such as Slack system events (channel renames, etc.). Slack
+        rejects the whole ``chat.postMessage`` call with
+        ``cannot_reply_to_message``; retrying the same payload as a
+        top-level channel message is safe and prevents the response from
+        being dropped entirely (#98285).
+        """
+        recoverable_codes = {"cannot_reply_to_message"}
         response = getattr(error, "response", None)
         response_get = getattr(response, "get", None)
         if callable(response_get):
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Symmetric fallback to existing Block Kit retry — non-threadable Slack roots (system events) that reject `thread_ts` with `cannot_reply_to_message` are now retried as top-level channel messages instead of dropping the response.

**What it does**
- `plugins/platforms/slack/adapter.py:10` in `send()` error path after `chat_postMessage`: `elif kwargs.get("thread_ts") and _is_thread_reply_rejection(e)` → clones `kwargs`, pops `thread_ts`/`reply_broadcast`, logs, retries `chat_postMessage(**retry_kwargs)`. Mirrors earlier `is_block_payload_rejection` structure.
- Adds `adapter.py:34` `_is_thread_reply_rejection(error)` — checks `error.response.get("error")=="cannot_reply_to_message"` or `str(error)` contains code (covers both SDK shapes).
- Tests `tests/gateway/test_slack_block_kit_adapter.py:67` — 3 cases: `cannot_reply_to_message` retries as top-level (count 2, first retains `thread_ts`/`reply_broadcast`/`blocks`/`text`, second drops thread keys and `blocks`, preserves `text`); `thread_not_found` still fails; no-thread target does not retry on same error.

**Non-blocking notes**
- Retry removes `blocks`? Second `retry_kwargs` is `dict(kwargs)` before `chat_postMessage` block fallback — if first attempt still had `blocks` (Block Kit rejection already handled earlier `elif`), second path is exclusive (`elif`), so this fallback posts `text`-only top-level — intentional per test `assert "blocks" not in second`; correct since `cannot_reply_to_message` is orthogonal to block rejection.
- Only `cannot_reply_to_message` is recoverable by dropping thread — other `thread_*` errors correctly surface.

liuhao1024:
Thanks for the review. Correct on the `elif` exclusivity — by the time this branch runs, the Block Kit rejection path has already been exhausted, so retrying as a text-only top-level message is the intended symmetric fallback. `cannot_reply_to_message` is the only thread error recoverable by dropping the thread keys; others still surface.

PR #98293, opened 2026-08-30
fix(update): shallow --check must not report "behind" when the compare API can't resolve a local-only HEAD (#98214)

### LINKED ISSUE #98214: update --check: shallow clone with local commits always reports "Update available" when the compare API can't resolve a local-only HEAD SHA
## Summary

On a **shallow** installer checkout that carries **local commits** (the maintained-custom-branch workflow that `updates.parked_branch_strategy: update_in_place` exists to support), `hermes update --check` reports

```
```

while the checkout is genuinely **up to date**: `origin/main` is an ancestor of `HEAD`, and `git rev-list --count HEAD..origin/main` is `0`.

The root cause is that `_github_compare_behind()` returns `None` (meaning *"could not determine"*) for a **local-only HEAD SHA that GitHub has never seen**, and the shallow branch of `_cmd_update_check()` has no case for `None` — it falls through to an `else` that asserts the checkout is behind.

**Unknown is reported as behind.**

## Affected code

`hermes_cli/update_cmd.py`, `_cmd_update_check()`, shallow fast-path. Observed at commit `4209d371a`, lines 4119-4149:

```python
if is_shallow:
    head_sha   = <rev-parse HEAD>
    target_sha = <rev-parse compare_branch>
    if head_sha and target_sha and head_sha == target_sha:
        print("✓ Already up to date.")
    else:
        from hermes_cli.banner import _github_compare_behind
        counted = _github_compare_behind(head_sha, target_sha)
        if counted == 0:
            print("✓ Already up to date.")
            return
        if counted is not None:
            print(f"⚕ Update available: {counted} {commits_word} behind {compare_branch}.")
        else:
            print(f"⚕ Update available (behind {compare_branch}).")   # ← None lands here
        print(f"  Run '{recommended_update_command()}' to install.")
    return
```

`counted == 0` is handled correctly. `counted is None` is not distinguished from "behind".

## Reproduction

1. Install via the installer so the checkout is shallow (`git clone --depth 1`).
2. Put local commits on top of the upstream tip — e.g. carry an unmerged fix on a custom branch, which is the documented `update_in_place` workflow:

```bash
git switch -c local-patches
git commit -m "fix: local patch not yet upstream"
```

3. Ensure the local commits have **not** been pushed anywhere GitHub can see them.
4. `hermes update --check`

## Observed vs expected

```
$ hermes update --check
→ Fetching from origin...
```

But the checkout is not behind:

```
$ git rev-parse --is-shallow-repository
true

0                       ← nothing to pull

$ git rev-list --count origin/main..HEAD
3                       ← our local commits

0                       ← origin/main IS an ancestor of HEAD
```

Direct probe of the helper, with the real SHAs:

```python
>>> from hermes_cli.banner import _github_compare_behind
>>> _github_compare_behind("8886ed41f3be...", "4209d371aa1b...")
None
```

`8886ed41f` is a local-only commit, so `GET /repos/.../compare/...` cannot resolve it and the helper correctly returns `None`. The caller then misreports that `None` as "behind".

Expected: `✓ Already up to date.`

## Root cause

The shallow path deliberately avoids `git rev-list HEAD..origin/main` because a shallow boundary can make the count meaningless, and delegates to the GitHub compare API instead — sound reasoning, since the remote graph is complete when the local one is truncated.

But that delegation has a blind spot: the API can only answer about commits **GitHub knows**. A local-only HEAD is exactly the case where it cannot, and `None` is then treated as a positive "behind" signal rather than as "no answer".

## Suggested fix

When the API cannot answer, fall back to a **local ancestry test**, which is valid on a shallow clone provided both refs are inside the shallow boundary (they are in this scenario — `origin/main` is `HEAD~3`):

```python
counted = _github_compare_behind(head_sha, target_sha)
if counted == 0:
    print("✓ Already up to date.")
    return
if counted is None:
    # The compare API cannot resolve a local-only HEAD. Ask git whether the
    # target is already contained in HEAD; exit 0 means we are ahead-or-equal,
    # never behind. Only report "unknown" if this is also inconclusive.
    ancestor = subprocess.run(
        git_cmd + ["merge-base", "--is-ancestor", compare_branch, "HEAD"],
        cwd=_m().PROJECT_ROOT, capture_output=True,
    )
    if ancestor.returncode == 0:
        print("✓ Already up to date.")
        return
    print(f"⚕ Update status unknown (could not compare against {compare_branch}).")
    print(f"  Run '{recommended_update_command()}' to install if you expect changes.")
    return
```

Two separable points, and the first matters more than the exact implementation:

1. **`None` must not mean "behind".** At minimum the wording should say the status is unknown rather than asserting an update exists.
2. `merge-base --is-ancestor` resolves the common case locally with no network, and is the same primitive the non-shallow path effectively relies on.

## Why this matters

- Every `--check` on such an install reports an update that does not exist, so the signal stops being trustworthy — and the banner nudges toward running `hermes update`.
- It specifically penalizes the **carried-local-commits** workflow the updater explicitly supports via `updates.parked_branch_strategy: update_in_place`. The state the feature is designed to produce is the state that triggers the false positive.
- Benign for a plain shallow clone with no local commits, since `head_sha == target_sha` short-circuits at the top.

## Related

- #93146 — same helper (`_github_compare_behind`), different failure: an *inflated* count via `banner.py` when HEAD is a merge commit with an old upstream parent. That report's suggested fix (compute locally with `merge-base` + `rev-list` instead of trusting the compare API) would also address this case, so the two are worth fixing together.

## Environment

- Hermes Agent v0.20.6 (2026.8.27), upstream `4209d371`
- Install method: git (installer, shallow)
- macOS 26.6.2, Python 3.11.16
- `origin` = `git@github.com:NousResearch/hermes-agent.git`, no `upstream` remote, so `compare_branch` = `origin/main`
- Checkout on a custom branch with 3 unmerged local commits, clean working tree

lEWFkRAD:
## What does this PR do?

Fixes the false "Update available" report from `hermes update --check` on a **shallow** checkout that carries **local-only commits** (#98214).

The shallow fast-path in `_cmd_update_check()` asks the GitHub compare API (`_github_compare_behind()`) to recover the behind-count. That API can only answer about commits GitHub has *seen* — for a local-only HEAD SHA (exactly the `updates.parked_branch_strategy: update_in_place` workflow the updater supports) it correctly returns `None` ("could not determine"), and the caller's `else` then *asserted* `⚕ Update available (behind origin/main)`. Unknown was reported as behind, forever, on a genuinely up-to-date checkout.

This PR:
1. **`None` no longer means "behind".** When the compare API has no answer, fall back to a local ancestry test: `git merge-base --is-ancestor <compare_branch> HEAD`. Exit 0 proves the target is contained in HEAD → `✓ Already up to date.` (no network needed).
2. **Honest wording for the residual case.** If git can't prove containment either (target outside the shallow boundary, or genuinely diverged), print `⚕ Update status unknown (could not compare against <ref>)` instead of claiming an update exists. The exit code and the `counted > 0` output bytes are unchanged, so script/cron gating on `update --check` keeps its current contract.

The fallback deliberately covers *both* reasons `_github_compare_behind` returns `None` — the local-only-SHA case *and* a failed/rate-limited API call — so offline shallow checks land on "unknown" or a real local answer rather than a fabricated "behind".

## Related Issue

Fixes #98214

## Type of Change

## Changes Made

- `hermes_cli/update_cmd.py` — `_cmd_update_check()`, shallow fast-path only (+17/−4): handle `counted is None` with a local `merge-base --is-ancestor` containment test; "already up to date" on rc 0, honest "status unknown" otherwise. The `head_sha == target_sha` short-circuit and the `counted == 0` / `counted > 0` branches are untouched (byte-identical output).
- `tests/hermes_cli/test_update_check_shallow_local_only.py` (new, 3 tests) — simulates the exact issue scenario (shallow repo, divergent SHAs, API → `None`) for all three ancestor outcomes: contained → "Already up to date."; rc 1 and rc 128 (inconclusive git error) → "unknown" wording, never "Update available".

## How to Test

1. Reproduce on unpatched main: shallow-clone the repo (`git clone --depth 1`), `git switch -c local-patches && git commit --allow-empty -m local`, ensure the commit is unpushed, run the shallow fast-path of `_cmd_update_check` with the compare API returning `None` → prints `⚕ Update available (behind …)` even though `git merge-base --is-ancestor origin/main HEAD` exits 0. (New tests are **red on main**: 3 failed.)
2. On this branch: `pytest tests/hermes_cli/test_update_check_shallow_local_only.py -q` → **3 passed**.
3. Regression: `pytest tests/hermes_cli/test_update_check.py tests/hermes_cli/test_update_apply_shallow_count.py "tests/hermes_cli/test_cmd_update.py::TestCmdUpdateCheckBranchFlag" -q` → **15 passed** (the CheckBranchFlag suite pins the non-shallow rev-list path, untouched by this diff).
4. Mechanism verified on a *real* shallow clone (git 2.53): local-only commit atop the tip → `merge-base --is-ancestor origin/main HEAD` exit **0**; target genuinely ahead → exit **1**; ref unresolvable → exit **128**. All three feed the right branch.
5. `scripts/check-windows-footguns.py --diff <merge-base>` → clean.

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate (no PR references #98214 at open time)
- [x] I've run the affected update/banner test modules (`tests/hermes_cli/test_update_check*.py`, `test_update_apply_shallow_count.py`, `test_cmd_update.py::TestCmdUpdateCheckBranchFlag`) — all pass; unrelated pre-existing npm-sandbox failures in `test_cmd_update.py` reproduce identically on the unpatched base (see below)
- [x] I've tested on my platform: Windows 11 (CPython 3.11, git 2.53, Git bash)

- [x] I've updated relevant documentation (README, `docs`, docstrings) — or N/A (no doc text asserted a "behind" guarantee; the change makes `--check` *more* truthful)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — `merge-base --is-ancestor` exit-code semantics are identical across platforms; the subprocess reuses the function's existing `git_cmd` (incl. the Windows `appendAtomically` flag) and `cwd`, with the same encoding/`errors="replace"` handling as its siblings

## Screenshots / Logs

Red on main (base `4209d371a`):
```
FAILED ...::test_none_with_ancestor_reports_up_to_date   (prints "⚕ Update available (behind upstream/main).")
FAILED ...::test_none_without_ancestor_does_not_claim_behind
FAILED ...::test_none_inconclusive_ancestor_does_not_claim_behind
3 failed in 1.20s
```
Green on branch: `3 passed in 1.80s`; combined targeted run `18 passed`.

Note for reviewers: in this repo's `test_cmd_update.py`, a few `TestCmdUpdateMigrationPrompt` / `TestCmdUpdateBranchFallback` tests fail identically on the unpatched base in my environment (they shell out to real `npm install` against the worktree and hit an npm versions/EPERM sandbox issue here). They do not touch the shallow `--check` path and are unrelated to this diff.

### DIFF
--- a/hermes_cli/update_cmd.py
+++ b/hermes_cli/update_cmd.py
@@ -4140,11 +4140,29 @@ def _cmd_update_check(branch: str = "main", *, branch_explicit: bool = False):
                 # Local commits on top of the remote tip — not behind.
                 print("✓ Already up to date.")
                 return
-            if counted is not None:
-                commits_word = "commit" if counted == 1 else "commits"
-                print(f"⚕ Update available: {counted} {commits_word} behind {compare_branch}.")
-            else:
-                print(f"⚕ Update available (behind {compare_branch}).")
+            if counted is None:
+                # The compare API has no answer. That covers both a local-only
+                # HEAD SHA GitHub has never seen (the update_in_place workflow)
+                # and a failed/rate-limited API call, so ask git locally
+                # whether the target is already contained in HEAD. On a shallow
+                # clone the answer is reliable while the boundary sits at or
+                # below the target ref (the local-commits-on-top-of-tip case);
+                # across a moved boundary git may still say no — that just
+                # lands on the honest "unknown" below, never a false "behind".
+                ancestor = subprocess.run(
+                    git_cmd + ["merge-base", "--is-ancestor", compare_branch, "HEAD"],
+                    cwd=_m().PROJECT_ROOT,
+                    capture_output=True,
+                    text=True, encoding="utf-8", errors="replace",
+                )
+                if ancestor.returncode == 0:
+                    print("✓ Already up to date.")
+                    return
+                print(f"⚕ Update status unknown (could not compare against {compare_branch}).")
+                print("  Run 'hermes update' to install regardless.")
+                return
+            commits_word = "commit" if counted == 1 else "commits"
+            print(f"⚕ Update available: {counted} {commits_word} behind {compare_branch}.")
             print(f"  Run '{recommended_update_command()}' to install.")
         return
 
--- a/tests/hermes_cli/test_update_check_shallow_local_only.py
+++ b/tests/hermes_cli/test_update_check_shallow_local_only.py
@@ -1,0 +1,79 @@
+"""RED tests for #98214 — shallow `update --check` misreports `None` as behind.
+
+Simulates the exact issue scenario: shallow clone, local-only HEAD commits on
+top of origin/main, GitHub compare API cannot resolve the local SHA and returns
+None. The shallow branch must NOT assert "Update available".
+"""
+
+import subprocess
+from types import SimpleNamespace
+from unittest.mock import patch
+
+import pytest
+
+
+def _shallow_side_effect(ancestor_rc=0):
+    """subprocess.run side effect: shallow repo, divergent SHAs, API -> None.
+
+    ancestor_rc: exit code of `git merge-base --is-ancestor <target> HEAD`
+      0  -> origin/main is contained in HEAD (ahead-or-equal, not behind)
+      1  -> genuinely not an ancestor
+      128 -> inconclusive (outside shallow boundary)
+    """
+
+    def side_effect(cmd, **kwargs):
+        joined = " ".join(str(c) for c in cmd)
+
+        if "rev-parse" in joined and "--is-shallow-repository" in joined:
+            return subprocess.CompletedProcess(cmd, 0, stdout="true\n", stderr="")
+        if "fetch" in joined:
+            return subprocess.CompletedProcess(cmd, 0, stdout="", stderr="")
+        if "rev-parse" in joined and "--abbrev-ref" in joined:
+            return subprocess.CompletedProcess(cmd, 0, stdout="local-patches\n", stderr="")
+        if "rev-parse" in joined and "--verify" in joined:
+            return subprocess.CompletedProcess(cmd, 0, stdout="", stderr="")
+        if "rev-parse" in joined and "HEAD" in joined:
+            return subprocess.CompletedProcess(cmd, 0, stdout="8" * 40 + "\n", stderr="")
+        if joined.rstrip().endswith("rev-parse origin/main"):
+            return subprocess.CompletedProcess(cmd, 0, stdout="4" * 40 + "\n", stderr="")
+        if "merge-base" in joined and "--is-ancestor" in joined:
+            return subprocess.CompletedProcess(cmd, ancestor_rc, stdout="", stderr="")
+        return subprocess.CompletedProcess(cmd, 0, stdout="", stderr="")
+
+    return side_effect
+
+
+ARGS = SimpleNamespace(check=True, branch=None)
+
+
+def _run_check():
+    from hermes_cli.main import cmd_update
+
+    with patch("hermes_cli.config.detect_install_method", return_value="git"), \
+         patch("hermes_cli.banner._github_compare_behind", return_value=None):
+        cmd_update(ARGS)
+
+
+def test_none_with_ancestor_reports_up_to_date(capsys):
+    with patch("subprocess.run", side_effect=_shallow_side_effect(ancestor_rc=0)):
+        _run_check()
+    out = capsys.readouterr().out
+    assert "Already up to date." in out, out
+    assert "Update available" not in out, out
+
+
+def test_none_without_ancestor_does_not_claim_behind(capsys):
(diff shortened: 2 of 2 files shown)


alt-glitch:

Related: #98167 already covers the local-ahead shallow-check mechanism across CLI, banner, and Desktop. This PR is the narrower CLI update-check variant.

PR #75441, opened 2026-07-31
fix(tts): retry transient Gemini failures

LuisMunozVillarreal:
## What does this PR do?

Gemini TTS transport failures could make gateway voice replies silently degrade to text-only delivery. Recent production logs exposed two related reliability gaps:

- long, one-shot Gemini synthesis requests repeatedly hit the 60-second read-inactivity deadline;
- the latest 2,707-character failure exhausted the old three-attempt budget with two read timeouts followed by a retryable HTTP 500.

This PR rebases the original retry fix onto current `main`, retries transient Gemini failures, bounds one-shot spoken-text size so long replies use the existing lossless chunking path, and adds a safe user-visible notice when gateway auto-TTS generation or delivery still fails.

## Type of Change

- [x] Bug fix (non-breaking change)
- [x] Documentation update
- [x] Tests

## Changes Made

- Retry Gemini TTS `requests.RequestException` failures, including read timeouts raised while consuming `stream=True` responses.
- Retry transient HTTP 408/429/500/502/503/504 responses while failing fast on deterministic client errors.
- Honor both delta-seconds and HTTP-date `Retry-After` values, bounded to one minute; malformed values fall back to the configured retry delay.
- Raise the default Gemini retry budget from 3 to 4 attempts, covering the observed timeout → timeout → HTTP 500 sequence while remaining configurable.
- Bound configuration to 1–10 attempts, 1–300 seconds of connect/read inactivity, and 0–60 seconds of retry delay; malformed and non-finite values fall back safely.
- Use a practical 2,000-character Gemini spoken-text cap. Longer replies are split through Hermes' existing ordered, sentence-aware chunking path without dropping text.
- Keep persona/performance direction outside the spoken-transcript chunk budget while independently bounding the composed Gemini prompt.
- Preserve bounded streamed-response reads and response cleanup.
- Avoid logging or displaying raw transport exception text, which may contain sensitive request details.
- Treat `success: false`, missing output files, unavailable provider requirements, thrown delivery exceptions, and unsuccessful `SendResult` values as auto-TTS failures.
- Preserve the original text reply, prepend a generic failure notice, and clean all temporary files after generation or partial delivery failure.
- Preserve current-main multi-file TTS delivery so chunked replies are delivered in order.
- Register and document the Gemini retry settings, practical chunk cap, real timeout semantics, and persona-prompt provider transmission.

## How to Test

Canonical targeted regression, gateway, and config coverage:

```bash
[...]

PR #98303, opened 2026-08-30
fix(tts): tolerate slow Gemini audio generation

arslanr-com:
## What does this PR do?

Gemini TTS goes through a single non-streaming `generateContent` call: the whole clip is synthesized server-side before the first byte comes back, so wall time scales with the length of the **audio**, not the request. The request timeout was hard-coded to `60`, which turns healthy long generations (~3–4k characters of spoken text) into failures:

```
ERROR tools.tts_tool: TTS generation failed (gemini): HTTPSConnectionPool(host='generativelanguage.googleapis.com', port=443): Read timed out. (read timeout=60)
WARNING agent.tool_executor: Tool text_to_speech returned error (76.35s): {"error": "TTS chunk 1 failed (gemini): ..."}
```

(Production log from a Telegram gateway deployment, Hermes 0.20.6 — the reply died at 76s while the synthesis itself was fine.)

Two changes, both bounded:

- the read timeout becomes configurable via `tts.gemini.timeout` (accepts `timeout_seconds` too, mirroring `_get_command_tts_timeout` semantics) and defaults to **300s**;
- **exactly one retry** (2 attempts total, linear backoff) covers read timeouts, 429 and 5xx. 4xx still fails fast — retrying a bad key or voice just burns another full synthesis wait. Opt out with `tts.gemini.retry: false`.

## Related Issue

No open issue found. Prior art: #75441 fixed the same gap plus two adjacent ones, went stale and was closed unmerged; its automated review confirmed the gap is still on `main` (`tools/tts_tool.py` makes a single request with no retry boundary). This PR is the narrow core of that fix — timeout + bounded retry — rebased on current `main`, with regression tests.

## Type of Change

## Changes Made

- `tools/tts_tool.py` — `_resolve_gemini_tts_timeout()`, `_gemini_tts_max_attempts()`, and the bounded retry loop around the `generateContent` POST in `_generate_gemini_tts()`
- `tests/tools/test_tts_gemini_timeout_retry.py` — 23 regression tests: default/override/invalid timeout resolution, retry on timeout/429/5xx, fail-fast on 4xx, opt-out, backoff bounds, response cleanup between attempts
- `cli-config.yaml.example` — the two new keys under `tts.gemini`

## How to Test

1. `pytest tests/tools/test_tts_gemini_timeout_retry.py -q` — 23 tests
2. Reproduce the bug: `tts.provider: gemini`, ask for a voice reply of ~3–4k characters. Before: `Read timed out. (read timeout=60)` after 60s. After: the same request completes (long clips legitimately take >60s server-side).
3. `pytest tests/ -q` — full suite.

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate (closest: #75441, closed unmerged; see above)
- [x] Tests: the new regression file passes (23/23, twice — on the fix's original base and rebased on current `main`). A full local `pytest tests/ -q` was started on the deployment box (2 vCPU / 4 GB) and interrupted for time after running without failures; relying on CI for the full matrix.
- [x] I've tested on my platform: Ubuntu 24.04 (Hetzner VPS, Telegram gateway deployment)

- [x] I've updated relevant documentation — `cli-config.yaml.example`; `docs/streaming-tts.md` covers the streaming path, which this change does not touch
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys
- [x] I've considered cross-platform impact (Windows, macOS) — `requests` timeout + `time.sleep` only, no platform-specific paths
- [x] I've updated tool descriptions/schemas if I changed tool behavior — N/A: new keys are `config.yaml` settings, not tool arguments; the tool's contract is unchanged

## Screenshots / Logs

Fix found and written against a live failure: a long Russian-language voice reply on a Telegram gateway died at the 60s read deadline while Gemini was still synthesizing. The failed call had already spent 76s wall time — the generation was proceeding server-side; only the client gave up.

### DIFF
--- a/cli-config.yaml.example
+++ b/cli-config.yaml.example
@@ -1382,6 +1382,8 @@ platform_toolsets:
 #     voice: "Kore"
 #     audio_tags: false
 #     persona_prompt_file: ""  # e.g. ~/.hermes/tts/radio-host.md
+#     timeout: 300          # read timeout in seconds; long clips synthesize server-side before the first byte
+#     retry: true           # one bounded retry on read timeout / 429 / 5xx; false fails fast
 #   xai:
 #     voice_id: "eve"       # built-in or custom voice ID from xAI Console
 #     language: "en"        # BCP-47 code ("en", "pt-BR") or "auto"
--- a/tests/tools/test_tts_gemini_timeout_retry.py
+++ b/tests/tools/test_tts_gemini_timeout_retry.py
@@ -1,0 +1,232 @@
+"""Regression tests for Gemini TTS request timeout and transient-error retry.
+
+Long-form Gemini TTS requests (~3-4k characters) routinely spend more than
+60s server-side before the single non-streaming ``generateContent`` response
+comes back, so the previously hard-coded ``timeout=60`` turned a healthy
+generation into ``TTS generation failed (gemini): ... Read timed out``.
+
+These tests pin the two behaviours that fix it:
+  * the read timeout is configurable via ``tts.gemini.timeout`` and defaults
+    to something well above 60s;
+  * exactly one bounded retry (with backoff) covers timeouts, 429 and 5xx,
+    while 4xx errors still fail fast.
+"""
+
+import base64
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+import requests
+
+
+@pytest.fixture(autouse=True)
+def clean_env(monkeypatch):
+    for key in (
+        "GEMINI_API_KEY",
+        "GOOGLE_API_KEY",
+        "GEMINI_BASE_URL",
+        "HERMES_SESSION_PLATFORM",
+    ):
+        monkeypatch.delenv(key, raising=False)
+    monkeypatch.setenv("GEMINI_API_KEY", "test-key")
+
+
+@pytest.fixture
+def fake_pcm_bytes():
+    # 0.1s of silence at 24kHz mono 16-bit = 4800 bytes
+    return b"\x00" * 4800
+
+
+def _ok_response(pcm: bytes) -> MagicMock:
+    resp = MagicMock()
+    resp.status_code = 200
+    resp.json.return_value = {
+        "candidates": [
+            {
+                "content": {
+                    "parts": [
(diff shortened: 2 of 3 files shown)


liyangbing:
<h3>Review</h3>
<p>A configurable read timeout plus a bounded retry is a focused repair for non-streaming Gemini TTS, and keeping 4xx failures fail-fast avoids retrying invalid credentials or voices. The main contract to make explicit is that each attempt owns its response and the total operation remains bounded.</p>
<h3>Suggested contract</h3>
<ul>
<li>Define the timeout and retry budget as one total operation deadline, including backoff, rather than allowing two full attempts to exceed the caller's tool or gateway deadline silently.</li>
<li>Retry only failures proven safe to retry—read timeout, 429, and selected 5xx—and preserve the original failure class when the final attempt fails.</li>
<li>Close or drain every response before the next attempt, avoid reusing a partial audio buffer, and publish the file or byte result only after the final response passes content and format validation.</li>
<li>Make the request idempotency story explicit. If the provider can synthesize after the client times out, a retry must not create an untracked duplicate charge or leave an orphaned audio artifact.</li>
<li>Keep configured timeout, attempt count, backoff, provider status, and elapsed time in redacted diagnostics; never log API keys, request bodies, or audio payloads.</li>
</ul>
<h3>Regression matrix</h3>
<ul>
<li>default timeout, valid override, invalid or negative override, and total-deadline exhaustion during backoff</li>
<li>read timeout, 429, 500, 503, 4xx credential or voice failure, malformed success response, and truncated audio</li>
<li>first attempt partial response, response cleanup, second-attempt success, both attempts failing, and retry opt-out</li>
<li>short and long text, concurrent TTS requests, cancellation during wait, process restart, and no orphaned temporary or audio files</li>
</ul>
<p>The handbook's <a href="https://sandbaseai.github.io/deepseek-harness-handbook/stuck-turn-stop-retry.html">bounded retry guide</a> documents the same deadline, failure-class, cleanup, and observable-outcome boundaries.</p>

PR #98305, opened 2026-08-30
fix(kanban): reject a review approval recorded by the implementer's profile

### LINKED ISSUE #98283: Kanban: review approval does not require an actor distinct from the implementer
# Kanban review flow: approval does not require an actor distinct from the implementer

Upstream report for NousResearch/hermes-agent. Measured on a live board, not inferred.

## Summary

`hermes kanban` records a review verdict as a `task_runs` row with
`outcome='review_approved'`. Nothing in the board machinery requires that row's
`profile` to differ from the profile that produced the work on the same card. A
worker can therefore request review and grant its own approval in-band, and the
card reaches `done` with an approval that is not independent.

This is a missing invariant, not a misconfiguration: there is no board setting
that enforces reviewer independence.

## Evidence

Board: `jocsans-website`, DB `~/.hermes/kanban/boards/jocsans-website/kanban.db`,
read-only. 20 cards carry at least one `review_approved` run. Seven of those
carry an approval whose profile had already done non-review work on the same card:

| card | approving run | profile | that profile's prior work on the card |
|---|---|---|---|
| t_f646e861 | 2953 | worker2 | 2950 timed_out, 2951 review_requested |
| t_70b34e22 | 2952 | worker2 | 2948 review_requested |
| t_a6622e92 | 2929 | worker2 | 2925 timed_out, 2928 review_requested |
| t_a9816345 | 2877 | critical-implementation | 2876 review_requested |
| t_bd97b281 | 2934 | critical-implementation | 2932 timed_out, 2933 review_requested |
| t_f3e4ceea | 2938 | critical-implementation | 2935 timed_out, 2937 review_requested |
| t_0b66ee0d | 2913 | reviewer | 2908 reclaimed |

Three distinct profiles appear as self-approvers (`worker2`,
`critical-implementation`, `reviewer`). This is a mechanism gap, not one
misbehaving profile.

The shape is consistently `review_requested` immediately followed by
`review_approved` under the same profile, with no intervening run by any other
actor.

## Why it matters

On this board a self-approval is the removal of the only control that catches
false completions. Concretely, `t_a9816345` was approved by the same profile that
implemented it, with approval metadata `mutation_proofs_verified: true` and
`consecutive_runs_verified: 3`, while its branch `wt/redos-font-normalize` had
**zero commits** and the entire change existed only as uncommitted worktree
changes. A `git worktree prune` would have destroyed the work while the board
read "reviewed and approved". Re-running the spec independently showed the
claimed "4/4 across 3 consecutive runs" was false: it failed 3/3.

Separately, `t_b676e1e7` closed `done` while an explicitly unmeasured item in its
own card body ("it may be a genuine reflow regression ... DO NOT assume;
measure") went unmeasured. The defect stayed live on the deployed origin for four
more days.

## Outcome audit: survivable only because a human intervened every time

Every one of the seven self-approved cards was audited against `main` on
2026-08-30 by measuring the artifact, not the run log. **All seven shipped.** No
work was lost. That is the honest headline, and it matters for severity.

But none of them were saved by the review that approved them. They were saved by
supervisor rescue commits and successor cards:

| card | how it actually landed |
|---|---|
| t_0b66ee0d | `wt/email-design-guards` merged as `1934f3a` |
| t_a9816345 | **reimplemented from scratch** on main as `7f4a581` |
| t_bd97b281 | `4ac01f4`, merged `0dcdd75` |
| t_f3e4ceea | `6356116`, merged `38d68fd` |
| t_f646e861 | scratch card; work landed via `2707c66` / `2ddd269` |
| t_70b34e22 | scratch card; work landed via `2707c66` |
| t_a6622e92 | **superseded** by successor t_bd97b281; its rescue commit was redundant |

Two cases show what the missing invariant actually costs:

- **t_a9816345** was approved with metadata `mutation_proofs_verified: true` and
  `consecutive_runs_verified: 3` while its branch had zero commits. The
  uncommitted worktree attempt used a boundary-dropping regex that diverged from
  the original on **15,619 of 230,000 fuzzed inputs**. It was discarded and
  rewritten, not merged. Had nobody looked, that approval would have shipped a
  broken regex or nothing.
- **t_a6622e92** needed a supervisor-authored rescue commit because its worker
  exhausted 150/150 iterations without committing, while a review run recorded
  `review_approved` for work that existed only in the worktree.

A control that holds only because a human independently re-checked all seven is
not a control. It is also worth noting that four of the seven closed `done` with
an **empty `result` field** while claiming success, which is a second, cheaper
tell available to the same machinery.

## Suggested invariant

Refuse to record `review_approved` when the approving `profile` matches any
profile that has a non-review run on the same task. Two viable placements:

1. Reject at write time in the review-approval path, with a clear error naming
   the conflicting run ids.
2. Allow the write but refuse the `done` transition, so the card parks in
   `review-required` until a distinct actor approves.

Placement (1) is preferable: it fails at the point of the untrue claim rather
than downstream.

A board-level opt-out may be warranted for single-profile boards where
independent review is not meaningful, but it should be explicit and off by
default.

## Local mitigation already in place

A read-only auditor (`review-integrity-gate.py`, supervisor profile, kept out of
the project repo per a standing ruling that host-specific agent guardrails do not
belong in shared repos) detects this shape from `task_runs` alone and flags all
seven instances above. It needs no separate reviewer card to compare against.

Verification of that auditor, both directions:

- positive control: all 7 known self-approvals flagged
- negative control: the other 13 approved cards are NOT flagged; each has an
  approver set disjoint from its worker set (e.g. `t_0b1ea52a` worker
  {critical-implementation, supervisor}, approver {reviewer})

So the gate is not simply flagging every approved card. It cannot, however, stop
the write — only report after the fact. That is why the durable fix belongs
upstream.

## A related shape that is NOT worth flagging

For completeness, since it looks tempting: a `completed` run of ~0 seconds
directly following a failed run on the same card. `t_b676e1e7` has exactly this
(runs: 2873 crashed, 2874 gave_up, 2875 completed in 0s).

Measured across the whole board it matches **32 cards**, most of them benign
bookkeeping where a supervisor closed a card after reclaiming a crashed run. As a
blocking rule it would be cry-wolf noise. Recorded here so the next person does
not spend the cycle: the duration column alone does not separate a false
completion from a legitimate administrative close.

Note for anyone reproducing: `task_runs.started_at` / `ended_at` are integer
epoch seconds, not ISO strings. A detector written against `datetime.fromisoformat`
silently matches nothing and reports a clean board.

chelsealong:
## What does this PR do?

Closes the four-eyes gap described in #98283: `hermes kanban`'s review lane
does not require the profile that approves a review to differ from any
profile that already did non-review (implementation) work on the same card.

`request_review()` leaves `assignee` pointed at the implementer when
`reviewer` is omitted on a first review. `claim_review_task()` then records
the *new* run's `profile` from that same `assignee`, so the implementer's own
profile ends up claiming the review lane. Nothing at approval time checked
this, so `complete_task()` would happily close the card — the exact
`review_requested` → `review_approved` self-approval shape the issue's board
audit found across seven cards on a live board, two of which shipped only
because a human happened to re-check the work by hand.

This PR adds the check at the point the approval is actually recorded
(`complete_task()`), which the issue calls out as the preferred placement:
it fails at the point of the untrue claim rather than downstream.

## Related Issue

Fixes #98283

## Type of Change

## Changes Made

- `hermes_cli/kanban_db.py`: new `SelfReviewApprovalError`, plus
  `_claimed_from_review()` and `_non_review_runs_for_profile()` helpers.
  `complete_task()` now raises `SelfReviewApprovalError` when the run being
  completed (`expected_run_id`) was claimed via `claim_review_task` (i.e. this
  is an actual review approval, not a plain implementation close — a review
  claim transitions `review -> running`, same status as a plain
  implementation claim, so `prior_status` alone can't distinguish the two;
  the run's own `claimed` event provenance can) and that run's profile
  already has a prior run on the card that was *not* claimed from review. A
  human closing a card with no active run (`expected_run_id=None`) is an
  explicit override and stays out of scope, mirroring `request_review`'s
  existing `force=` escape hatch.
- `hermes_cli/kanban.py`: `_cmd_complete` catches the new error and prints a
  clean diagnostic instead of an uncaught traceback (a worker completing its
  own review claim via the CLI, not just the `kanban_complete` tool, can hit
  this path).
- `tools/kanban_tools.py`: `kanban_complete` catches the new error with a
  worker-facing message telling it to route the card to a distinct reviewer,
  matching the existing `HallucinatedCardsError`/`ArtifactPreservationError`
  handling pattern in the same function.
- `tests/hermes_cli/test_kanban_review_lifecycle_complete.py`: new
  `test_complete_task_rejects_self_review_approval`, reproducing the exact
  self-approval shape from the issue (reviewer omitted → same profile claims
  review → same profile calls `complete_task`) and asserting it's rejected
  without mutating the card.
- `tests/plugins/test_kanban_dashboard_plugin.py`: one existing test
  (`test_reopening_parent_retracts_review_and_blocks_approval`) happened to
  use the same profile name for the card's implementer and its review
  approval as incidental test data (unrelated to what the test actually
  verifies — parent-reopen retraction). Added an explicit distinct
  `reviewer="approver"` so it keeps testing parent-reopen behavior instead of
  incidentally re-exercising the self-review loophole this PR closes.

## How to Test

1. `scripts/run_tests.sh tests/hermes_cli/test_kanban_review_lifecycle_complete.py -q`
2. Confirm the new test fails without the fix:
   `git checkout HEAD~1 -- hermes_cli/kanban_db.py hermes_cli/kanban.py tools/kanban_tools.py`,
   re-run the same test (`AttributeError: module 'hermes_cli.kanban_db' has no
   attribute 'SelfReviewApprovalError'` on the fully-reverted source — the
   invariant does not exist on `main` today), then restore.
3. `bash scripts/run_tests.sh tests/hermes_cli/ -q` — 743 files, 6192 passed,
   2 failed, 71 skipped. Both failures are pre-existing and reproduce
   identically on unmodified `main` (`test_kanban_review_surfaces.py::test_review_tools_are_gated_and_visible_to_kanban_workers`
   and `test_linux_desktop_entry.py::test_exec_leaves_venv_shebang_scripts_alone`,
   the latter an interpreter-path artifact of this sandbox, unrelated to
   kanban).
4. `ruff check` and `ty check` on all touched files: clean / only the same 19
   pre-existing `ty` diagnostics present on unmodified `main` (none in the
   touched line ranges).

### Code

- [x] I searched for existing PRs to make sure this isn't a duplicate — closest
      is #97933, which guards *reviewer assignment* at `request_review()` time
      (rejecting a first review when `reviewer` is omitted and would default
      to the implementer). It's complementary, not overlapping: it doesn't
      touch the approval path, so a reviewer assigned before that guard
      existed, an explicit-but-conflicting reviewer, or a re-review whose
      provenance recovers a profile that also did earlier implementation work
      on the card would still slip through it. This PR closes the gap at the
      point the approval itself is recorded.
- [x] I've run `pytest tests/ -q` (via `scripts/run_tests.sh`) and all tests
      pass except two pre-existing, unrelated failures (see above)

- [x] N/A — no config keys, no architecture/workflow changes, no tool
      schema changes (the `kanban_complete` tool's *error message* changed,
      not its schema)

## AI assistance disclosure

This PR was written with AI assistance (Claude Code / Claude Sonnet 5),
operating autonomously against the issue text and the current `main` branch,
under human-configured guardrails. The diagnosis, fix design, and test were
verified by running the real test suite, `ruff`, and `ty` against this
checkout before pushing.

### DIFF
--- a/hermes_cli/kanban.py
+++ b/hermes_cli/kanban.py
@@ -2392,13 +2392,19 @@ def _cmd_complete(args: argparse.Namespace) -> int:
                 failed.append(tid)
                 continue
 
-            if not kb.complete_task(
-                conn, tid,
-                result=args.result,
-                summary=summary,
-                metadata=metadata,
-                expected_run_id=_worker_run_id_for(tid),
-            ):
+            try:
+                completed = kb.complete_task(
+                    conn, tid,
+                    result=args.result,
+                    summary=summary,
+                    metadata=metadata,
+                    expected_run_id=_worker_run_id_for(tid),
+                )
+            except kb.SelfReviewApprovalError as self_review_err:
+                failed.append(tid)
+                print(f"kanban: {self_review_err}", file=sys.stderr)
+                continue
+            if not completed:
                 failed.append(tid)
                 print(f"cannot complete {tid} (unknown id or terminal state)", file=sys.stderr)
             else:
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -5360,6 +5360,71 @@ class ArtifactPreservationError(RuntimeError):
     """Raised when a declared scratch deliverable cannot be preserved."""
 
 
+class SelfReviewApprovalError(ValueError):
+    """Raised by ``complete_task`` when the profile recording a review
+    approval already has a non-review run on the same card (#98283).
+
+    Four-eyes review is meaningless if the approving profile is the same
+    one that did (or attempted) the implementation work being approved.
+    ``conflicting_run_ids`` carries the offending prior run ids for the
+    caller to surface in a diagnostic.
+    """
+
+    def __init__(self, task_id: str, profile: str, conflicting_run_ids: list[int]):
+        self.task_id = task_id
+        self.profile = profile
+        self.conflicting_run_ids = list(conflicting_run_ids)
+        super().__init__(
+            f"review approval blocked: profile {profile!r} already has "
+            f"non-review run(s) {conflicting_run_ids} on {task_id}; a "
+            f"distinct profile must record the approval"
+        )
+
+
+def _claimed_from_review(conn: sqlite3.Connection, task_id: str, run_id: int) -> bool:
+    """True if ``run_id`` was claimed via :func:`claim_review_task`.
+
+    Determined from the run's own ``claimed`` event payload, the same
+    provenance :func:`request_changes` already relies on to confirm a
+    rejection came from an active review claim.
+    """
+    row = conn.execute(
+        "SELECT payload FROM task_events "
+        "WHERE task_id = ? AND run_id = ? AND kind = 'claimed' "
+        "ORDER BY id DESC LIMIT 1",
+        (task_id, int(run_id)),
+    ).fetchone()
+    try:
+        payload = json.loads(row["payload"]) if row and row["payload"] else {}
+    except (json.JSONDecodeError, TypeError):
+        payload = {}
+    return isinstance(payload, dict) and payload.get("source_status") == "review"
+
+
+def _non_review_runs_for_profile(
+    conn: sqlite3.Connection,
+    task_id: str,
+    profile: str,
+    *,
+    exclude_run_id: int,
+) -> list[int]:
+    """Prior run ids on ``task_id`` by ``profile`` that were NOT claimed
+    from review — i.e. implementation work, not review work.
+
+    Used to enforce that the profile approving a review is distinct from
+    any profile that already touched the card as an implementer.
+    """
+    rows = conn.execute(
+        "SELECT id FROM task_runs WHERE task_id = ? AND profile = ? AND id != ?",
+        (task_id, profile, int(exclude_run_id)),
+    ).fetchall()
+    return [
+        int(row["id"])
+        for row in rows
+        if not _claimed_from_review(conn, task_id, row["id"])
+    ]
+
+
 def complete_task(
     conn: sqlite3.Connection,
     task_id: str,
@@ -5450,6 +5515,34 @@ def complete_task(
             (task_id,),
         ).fetchone()
         prior_status = prior["status"] if prior else None
+        # Four-eyes guard (#98283): a review approval recorded by the same
+        # profile that already did non-review (implementation) work on this
+        # card is not an independent review. A review claim transitions
+        # review -> running (claim_review_task), same as a plain
+        # implementation claim, so ``prior_status`` alone can't tell the two
+        # apart here — the run's own ``claimed`` event provenance can. Only
+        # checkable when the caller proves which run is approving
+        # (expected_run_id); a human closing a card with no active run
+        # (expected_run_id=None) is an explicit override and stays out of
+        # scope, same as request_review's force=.
+        if (
+            expected_run_id is not None
+            and _claimed_from_review(conn, task_id, int(expected_run_id))
+        ):
+            approver_row = conn.execute(
+                "SELECT profile FROM task_runs WHERE id = ?",
+                (int(expected_run_id),),
+            ).fetchone()
+            approver_profile = approver_row["profile"] if approver_row else None
+            if approver_profile:
+                conflicts = _non_review_runs_for_profile(
+                    conn, task_id, approver_profile,
+                    exclude_run_id=int(expected_run_id),
+                )
+                if conflicts:
+                    raise SelfReviewApprovalError(
+                        task_id, approver_profile, conflicts,
+                    )
         if expected_run_id is None:
             cur = conn.execute(
                 """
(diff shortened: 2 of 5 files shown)


Enough1122:

**Overall:** Enforces four-eyes review by rejecting self-approval at kanban complete.

**What it does**
- `hermes_cli/kanban_db.py:SelfReviewApprovalError` + helpers `_claimed_from_review` / `_non_review_runs_for_profile` use claimed-event provenance (`source_status=="review"`) to distinguish review claims from implementation claims.
- `complete_task` now raises `SelfReviewApprovalError` when `expected_run_id` is a review claim whose profile already has non-review runs on same card. Human close (`expected_run_id is None`) remains override.
- `hermes_cli/kanban.py:_cmd_complete` and `tools/kanban_tools.py:_handle_complete` surface error instead of silently completing.
- Test pins same-profile implement->review->approve blocked and card stays `running`; existing test updated to pass explicit `reviewer="approver"`.

**Non-blocking notes**
- `_claimed_from_review` treats missing/invalid JSON as non-review (fail-open for provenance) — reasonable, but a malformed claimed payload would bypass the guard; a debug log could help audit.
- Query does `SELECT id FROM task_runs WHERE profile=?` then filters per-row with extra query; N+1 is fine for typical few runs but worth noting if a card accumulates many runs.

No functional issues found.

PR #98073, opened 2026-08-29
feat(groups): control Group Chats from messaging

dokterdok:
## Control Group Chats from messaging

When away from Desktop, people need a way to check a Group Chat's progress and give direction from their messaging app. This PR supplies `/group` to find a group, see its Bots and recent messages, send a follow-up, retry unfinished work or Stop; authorized owners can also inspect exact pending approvals and remember an eligible command for one Bot in one Group Chat.

**A companion to Desktop, focused on quick check-ins.** Messaging is for seeing progress, giving direction, handling approvals or getting a result while on the go. It intentionally does not reproduce every Group Chat feature inside another messaging app. Desktop remains the primary experience today; a future native Hermes app for iOS or Android would be the natural place for a fuller, more familiar mobile group-chat experience.

This PR supplies the **shared command and authorization core, with the backend and classic-control support it consumes**. Hosted and Desktop-driven groups use their existing execution paths; classic groups still need Desktop for execution and approvals. The richer native Group Chat/Files consumer is composed separately in #98307.

## Fewer Repeat Approvals

**Always allow in this chat** asks the owner to confirm the exact command and its connection/folder before remembering it. `/group N permissions` shows saved entries; `forget` removes one. Other Bots/groups do not inherit the permission and protected denials remain in force. Text-only clients have the same warning, confirmation and removal commands.

The authored permission increment remains separately selectable at `3a74e903b2`. Permission labels and expired requests were clarified at `322ee31472`; no profile-wide or gateway-wide approval bypass is enabled.

## Example

In an authorized Signal chat:

```text
/group
/group 1
/group 1 bots
/group 1 send @writer Summarize the latest decision in three bullets.
```

Use the number and handle shown in your own list. Telegram's composed menus offer the same actions; Slack and Matrix use `!group`. Shared Home chats require explicit audience consent, and `/sethome` alone does not grant authority. Stop and exact decisions retain their authorization and replay fences during busy work.

<details>
<summary>Complete command reference, including the optional Files layer</summary>

Replace `N` with the stable group number from **your** list. Values in angle brackets are placeholders. Bot numbers belong to the displayed roster; approval, permission and file codes are the exact codes shown by Hermes, not list positions.

**Included in this PR**

| Command | What it does |
| --- | --- |
| `/group` | Open the group picker, or show the first list page. |
| `/group help` | Show short help. |
| `/group list [page]` | List all accessible groups, eight per page; start at page 1. |
| `/group N` | Show status, participating Bots and recent message previews. |
[...]

PR #98072, opened 2026-08-29
feat(groups): share Group Chat files across gateways

dokterdok:
## Keep shared Group Chat files available across gateways

If a user shares a file in a Group Chat, every authorized Bot should receive the exact accepted bytes even when it runs on another gateway. A restart, retry, owner replacement, or later read must not substitute a different version, lose the file reference, or leak bytes outside the frozen room and recipient authority.

This draft implements the bounded Files/input slice: canonical upload, listing, search, download and exact accepted-input replay for authorized Group Chat participants across local and cross-profile handling. It does not enable unsupported target modes, standalone receiver bootstrap, native prepared Files, Bot-produced output, or live deployment; the required lower-layer owners and remaining qualification gates are listed below.

## What this PR does

- Uploads, lists, searches, and downloads retained Group Chat files through the canonical Files handlers and viewer catalog.
- Keeps event-, room-, authority-, version-, and recipient-bound references across canonical local/cross-profile input handling.
- Retains signed peer-input staging, exact accepted-input replay, bounded batch planning, revocation cleanup, legacy local delivery, and named-profile receiver tests without enabling unsupported target modes.
- Restores the selective authority-claim index during schema repair and requires both Files lookup indexes before treating the schema as current. The steady-state lookup is bounded; one-time index construction and large-store bootstrap are not claimed bounded.

### Important protections

- **P1 — file access and lifetime:** recheck the captured owner after I/O; preserve immutable accepted inputs and recipient authorization. A replaced owner must not receive bytes selected under the previous owner.
- **P1 — stale authority and refused turns:** retain authority gateway/epoch isolation in spool and grant operations. Proven nonadmission rolls back legacy staging so a later text-only turn cannot inherit rejected attachments; an ambiguous submission remains ambiguous.
- **P2 — revocation during staging:** hold every enforcing grant store behind a write fence through spool commit, including the shared-store failure path. If a profile fence cannot be acquired, refuse the mutation.
- **P2 — duplicate work and oversized input:** resolve the complete recipient-bound file batch before upload, enforce the aggregate budget before capture, and use the existing exact replay receipt rather than rematerializing accepted input.

These labels describe the impact of the addressed failure classes, not maintainer-assigned priorities or newly established upstream blockers.

## Review shape and history

[...]

PR #98307, opened 2026-08-30
chore(groups): compose Group Chat layers in a permanent integration draft

dokterdok:
## Test Group Chat layers together in a permanent draft

A Group Chat can cross Desktop, gateways and messaging, so checks of one layer alone cannot show whether the combined experience works. This permanent draft brings selected layer and fix revisions together for integration testing and UAT; the owning PRs remain the delivery path, and this PR is not intended to land.

## Runtime refresh

This moves the existing #97681 integration forward onto #106742's canonical runtime. This refresh merges runtime `b36398f71929` into the previously published #98307 head `0860dc1feadc`. Normal merges preserve the authored history rather than replacing the branch with a new snapshot.

Current head: [`cb657c411cda`](https://github.com/dokterdok/hermes-agent/commit/cb657c411cda8a3b645e9cbfa2a2c8df5ef9d4c5). Tested runtime base: [`b36398f71929`](https://github.com/NousResearch/hermes-agent/commit/b36398f71929d6069905ccca59af8d1516c5ee3a).

**This PR stays permanently draft and is not intended to land.** It is the integration/UAT view for selected layer and fix revisions; the owning PRs remain the delivery path. Those PRs should contain only layer-owned changes and declare their dependencies, rather than absorb this composition or each lower layer. The required-versus-recommended fix map lives in #97681. A useful UAT fix is not automatically a hard dependency, and inclusion here does not establish native/recovery acceptance.

## What this update reconciles

| Area | Change and boundary |
| --- | --- |
| Runtime ownership | Canonical `SessionAuthority` remains the executor. Canonical, stale or unavailable capabilities cannot silently select a legacy Desktop executor. |
| Desktop UX integration | Retains same-ID snapshot/queued-transcript protection and durable Send journals while adopting the shared `JsonRpcRequestChannel`. An open transport is not automatically completed replay or settled work. |
| Files | Integrates canonical catalog/version-bound reads and the corrected download sink. Atomic accepted-input custody is preserved alongside upstream ranged reads and digest-attested preflight. Missing or corrupted retained bytes refuse execution without another transfer; upstream physical-alias retention is kept. |
| Uncertain work | Observer failure/cancellation cannot manufacture terminal Run state or expire its original retry identity. Retained peer attempts use receipt-only observation; missing receipts stay unresolved. |
| Pending controls | Unsupported prompts retain their target identity through storage, messaging and Desktop. They are displayed as observations and cannot queue a decision. |
| Current runtime fixes | Keeps the full upstream deletion fence: deleted sessions cannot reappear through late backfill, and busy/unknown work refuses the entire delegate batch. Failed pre-admission seeds can be recreated without a deletion tombstone. Frozen-worker config isolation is retained. |
| Adapter and test ownership | Carries per-profile settings into the extracted adapter owners: explicit scoped environment, then the profile’s own YAML, then defaults. Native choice/file/source guards and upstream test/docs relocations remain intact. |

New canonical RoomLink admissions and peer control mutations remain explicitly unavailable where the required contracts are not implemented. Named-peer readiness remains OFF. Retained observations do not grant execution permission, and the supported standalone/legacy paths are not an automatic fallback for canonical failures.

## Verification on this update

Fresh checks on the `b36398f71929` composition:

| Scope | Passed | Qualification |
| --- | ---: | --- |
| Backend, Files/custody, config and native-media alias handling | 148 | 18 selected files; temporary data, no live turns. |
| Maintenance and caller behavior | 91 | Three selected files; busy/unknown refusal and batch atomicity. |
| Upstream deletion, seed discard and sweep regressions | 12 | Original upstream assertions retained; reset/adoption cases excluded. |
| Scoped adapter settings and native-choice components | 68 | Five selected files; mocked transports, not live messaging. |
| Desktop UI | 243 | Seven selected files; hydration, ownership, draft recovery and Files. |

**319 selected Python cases and 243 UI cases passed**, with the canonical Python runner and zero retries. Desktop typechecking and scoped Python lint passed. The renderer build passed; all 126 entry-script/stylesheet/preload references resolve to nonempty assets. These are bounded checks, not full-suite or native acceptance.

The retained-input regression covers corrupted and missing v3/legacy copies without retransfers or SQL changes. Maintenance assertions now require the upstream retirement tombstone; seed compensation remains distinct from explicit deletion. Earlier shared-channel, journal and field receipts stay attached to their tested revisions and are not counted again here.

The PR residual passes whitespace checks against this runtime base; inherited base diagnostics were not reformatted as unrelated work.

## Remaining gates

- No matching-build native/Mac/Windows/fleet or full-repository release acceptance is claimed for this merge. Historical field receipts belong to their original revisions.
- Participant emergency Stop, fault/source-closing/unknown-resolution, paused recovery, exclusivity and physical failover remain separate held work. This does not replace their original review authority.
- This update does not complete native New Bot installation-management, Group artwork or all API artifact/display contracts, and does not expand permissions to bypass them.
- Remote SSH/URL/Cloud/standalone-Serve cutover follows the runtime's separate contracts; local canonical evidence is not remote parity.
- GitHub's exact-head CI check suites require maintainer approval (`action_required`); no hosted checks have executed. Local passes are not CI approval, and no repository settings or workflows were changed to bypass it.

<details>
<summary>Owning source PRs and preserved contributions</summary>

| Source | Responsibility |
| --- | --- |
| #99107 | Passive-copy namespace/quarantine and execution safety. |
| #99960 / #100016 | Admission uncertainty, routes, grants and retirement. |
| #97846 | Desktop UX integration and source-bound client state. |
| #98072 / #99159 | Shared-input Files and producer/publication ownership. |
| #104198 / #104199 | Classic producer exports and the Desktop Files consumer. |
| #98073 | Messaging commands, authorization, Home consent and scoped approvals. |
| #104601 | Passive history and unfinished-work evidence; #104759 remains an absorbed donor. |
| #105079 / #105197 | Guarded recovery carrier and retained inert donor; no new activation is implied. |
| #103638 / #103214 / #100755 / #105467 | Policy scope, SQLite/config support and mention/command parsing. |
| #109338 | Generic runtime, input-custody and journal prerequisites. Upstream has absorbed selected fixes, including native-media alias retention, not the entire deferred batch. |

The original #98307 ancestry retains @frizikk's #105984/#105985, Jugol's #96162, bennybuoy's #97914 and Anhtahaylove's #101478 contributions. Cross-gateway foundations retain work from @pesho-vsn (#88408), @giaiant (#94336), @pierrenode (#88819) and @liuhao1024 (#93952). #104198/#104199 retain the sources first published as fork PRs #2/#1. Selected Files ports preserve author, date, original message and source trailers.

Historical native/field evidence remains available through the previous published revision and its recorded receipts. It is not relabeled as acceptance of the new runtime composition. #97681 remains the full layered programme; publishing this draft checkpoint does not finish that programme.

</details>


liyangbing:
<h3>Review: add crash-window tests around artifact settlement</h3><p>The integration description says upload, commit, acknowledgement, redelivery, restart, and cleanup are idempotent. In the hosted attachment path, those claims cross several different durability boundaries: <code>put()</code> writes the blob before its SQLite row commits; <code>commit_message_with_receipt()</code> changes an uploaded row to committed; <code>retain_event()</code> removes the expiry only after the room event is durable; and <code>prune()</code> commits row/blob metadata deletion before unlinking the filesystem object, with orphan sweeping as recovery.</p><h3>Suggested focused matrix</h3><p>Inject a crash or forced exception at each boundary and restart the store/room owner:<br>1. after the blob write but before the attachment transaction commits — the orphan must be reclaimable and must not bypass the gateway quota on the next upload;<br>2. after the attachment is committed but before the immutable room event is durable — the expiry/abort path must not retain private bytes;<br>3. after the event is durable but before <code>retain_event()</code> — reconciliation must retain exactly the IDs named by that event, including a repeated recovery call;<br>4. after <code>prune()</code> commits metadata deletion but before filesystem unlink — restart cleanup must remove the orphan without deleting a newly-created blob;<br>5. during duplicate commit/ack/redelivery from two workers — one logical receipt should advance once, while the final read remains authorized to the intended recipient only.</p><p>For every row, assert the SQLite state, blob digest/size, room and gateway quota counters, visible event payload, and the next cold-start read. The random <code>blob_id</code> and digest checks make this feasible without snapshotting implementation details. This would turn the draft's strong idempotency claims into regression evidence at the exact crash windows where database durability and filesystem durability can diverge.</p>

dokterdok:
Addressed in `53c72affc4`.

The focused matrix now injects and verifies:

- blob write before attachment-row commit, including cold orphan cleanup and next-upload quota;
- attachment commit before room-event publication, including expiry/abort cleanup;
- room event before `retain_event()`, including repeated reconciliation;
- prune metadata commit before filesystem unlink, including restart cleanup without deleting a newly-created live blob;
- concurrent duplicate commit/retain workers, quota counters, visible payload, and recipient-only reads.

`retain_event()` now changes only still-expiring rows, so duplicate workers advance the logical receipt once instead of merely converging to the same bytes. The complete focused attachment/classic/policy matrix passes **80/80**.

The subsequent packaged UAT also exercised image, PDF, generic-file, Bot-generated-file, close/reopen, and cross-gateway handoff paths on real gateways. Thanks for pointing the review at the filesystem/SQLite boundaries rather than only the happy-path RPCs.

andrexibiza (commented):
Reviewed exact integration head `757f804ccfa5f9e6b749fe30ee8bc38e65986dbc` as the evidence build, not as the intended merge path. Exact-head CI, Docker, and Nix are green, and the combined UAT/crash-window evidence remains useful.

The integration result cannot supersede unresolved source-layer defects. This tree currently composes all four:

1. #97712: replay pages enforce `MAX_LOG_PAGE_BYTES` with SQLite character counts rather than UTF-8 byte counts.
2. #97797: near-expiry grant renewal silently adopts a changed execution policy/capability catalog and rewrites the pending dispatch instead of requiring explicit reauthorization.
3. #97846: transient hosted-worker-unavailable RPC code `4123` is classified as terminal, permanently failing persisted commands that should replay after worker recovery.
4. #98072: RoomLink attachment spool identity and cleanup omit authority gateway/epoch, allowing stale-lineage collision and old-grant cleanup to delete current-epoch batches.

The evidence tree also inherits the touched-file violations of the repository's hard 2,000-physical-line ceiling, including production godfiles and the new Messaging test godfile.

None of the reported green/UAT matrices exercises the exact adverse shapes above. Keep this branch as an integration harness, but it is not acceptance evidence for the six source heads until those defects are repaired in their owning layers, recomposed here without flattening authorship/dependency boundaries, and re-proven on the new exact object.

dokterdok:
The only red check at 5ac66c8dd0 is the CI-review gate. The workflow diff adds `--extra messaging` to the two existing dependency installs in `.github/workflows/tests.yml`; triggers, permissions and test assertions are unchanged. All executed technical checks passed. Could a maintainer review that two-line workflow change and apply `ci-reviewed` if satisfied?

dokterdok:
One CI follow-up on `97127c64873`: the unchanged `session-unread-tile` test hit its 15-second timeout; the other 8,137 UI tests passed. All three cases pass locally at the exact CI merge, without changing the timeout. Could a maintainer rerun just the [JS job](https://github.com/NousResearch/hermes-agent/actions/runs/34081817951/job/101618436061)? GitHub refused my rerun request for lack of repository permissions. The separate workflow-review gate is also unchanged.

tui_gateway/hosted_room_replication.py
```diff
@@ -0,0 +1,607 @@
...
+                    return self._http_failure(route, checkpoint, exc)
+                if not retirement.confirm_home_enrollment(
+                    self.db_path, enrollment_id=enrollment["enrollment_id"], proof=proof,
+                ):
+                    return False
+            elif enrollment["state"] != "enrolled":
+                return False
+        cursor, pending = checkpoint["acked_seq"], checkpoint["pending_end"]
+        if pending is None and cursor >= route.room["latest_seq"] and checkpoint["status"] == "acked":
+            return self._publish_work_records(route, checkpoint, client)
```
andrexibiza: **[P2, inherited from #104759] Compose the source-owned work-record scheduling fix here.** This entire publisher blob matches `6eca08e21fa5` (`ae6e7419fb50050c36d9d1c5962f19c77db8d01b`), including the quiet-history gate. The [source review](https://github.com/NousResearch/hermes-agent/pull/104759#pullrequestreview-5129749535) reproduces 100 successful history turns with no work-record calls, followed by a successful quiet-turn control. Preserve a consistent snapshot/history anchor and bounded record-delivery opportunity; do not relax prefix or exact-ACK checks. Repair #104759 and compose that authored commit, rather than introducing an integration-only scheduler variant.

gateway/hosted_room_replica_retirement.py
```diff
@@ -0,0 +1,898 @@
...
+            BEFORE INSERT ON {table}
+            WHEN EXISTS (SELECT 1 FROM {RETIREMENT_TABLE} WHERE room_id=NEW.room_id)
+            BEGIN SELECT RAISE(ABORT, 'replica copy is retired'); END""")
+    conn.execute(f"""CREATE TRIGGER IF NOT EXISTS trg_replica_retired_event_update
+        BEFORE UPDATE ON hosted_room_replica_events
+        WHEN EXISTS (SELECT 1 FROM {RETIREMENT_TABLE} WHERE room_id=OLD.room_id)
+        BEGIN SELECT RAISE(ABORT, 'replica copy is retired'); END""")
+    conn.execute(f"""CREATE TRIGGER IF NOT EXISTS trg_replica_retired_room_update
+        BEFORE UPDATE ON hosted_room_replicas
+        WHEN EXISTS (SELECT 1 FROM {RETIREMENT_TABLE} WHERE room_id=OLD.room_id)
```
andrexibiza: **[P2, inherited from #104601] The new work-record guards do not protect history UPDATE destinations.** These two history/replica predicates still inspect only `OLD.room_id`. The `initialize_retirement_guards()` call below installs guards on the work-record table; it does not prevent moving an active history/replica row into a retired destination. See the [source review and bounded SQLite proof](https://github.com/NousResearch/hermes-agent/pull/104601#pullrequestreview-5129814792). Compose the source-owned fix, including migration of existing trigger definitions and allowed-bookkeeping controls. This is the same storage-invariant finding, not a separate API exploit or a request for duplicate implementation here.

andrexibiza (commented):
## Focused composition review — history, work records and retirement

Reviewed the composed publisher and retirement guards at `04979b1749934e3a9bef4568bc0db8e00f32314e`. This is a bounded source-composition review, not a review or acceptance of the entire 446-file integration diff.

Two findings from the owning source reviews survive in this composition:

1. **#104759 — P2 work-record starvation during ongoing history traffic.** [Source review and reproduction](https://github.com/NousResearch/hermes-agent/pull/104759#pullrequestreview-5129749535). `tui_gateway/hosted_room_replication.py` has the same Git blob, `ae6e7419fb50050c36d9d1c5962f19c77db8d01b`, in source `6eca08e21fa5b2c500fbd0998bbd936e64db6cd8` and this integration head. The source's history-first gate and final-page return are therefore unchanged here. A healthy target can acknowledge continuous history without receiving task-record snapshots until a quiet publisher turn occurs.

2. **#104601 — P2 storage-guard destination omission.** [Source review and SQLite reproduction](https://github.com/NousResearch/hermes-agent/pull/104601#pullrequestreview-5129814792). The composed retirement owner, blob `d6df387b56150123f496146f697b619bda38c1f2`, retains the two `OLD.room_id`-only UPDATE guards. Its added call to `initialize_retirement_guards()` protects the work-record target table, not `hosted_room_replicas` or `hosted_room_replica_events`; it does not close the history-row destination gap. This remains a raw-SQL invariant finding, not evidence of an HTTP authentication bypass or successor execution.

Keep the fixes in their authored sources: retirement guard/migration correction in #104601; bounded snapshot/anchor delivery fairness in #104759. Compose those commits here rather than adding a second implementation. The source reviews contain the required regressions and the precise limits of the isolated probes; their counts are not new integration or native-client test runs.

For this integration head, [CI is failed](https://github.com/NousResearch/hermes-agent/actions/runs/34093373915), [Docker succeeded](https://github.com/NousResearch/hermes-agent/actions/runs/34093373123), and [Nix succeeded](https://github.com/NousResearch/hermes-agent/actions/runs/34093373137). I have not independently diagnosed every CI job in this review, and am not treating the aggregate as an all-green release receipt. Current-head checks likewise do not certify every surviving commit.

The current ownership table already describes #104198 as temporarily draft, so the earlier description mismatch is no longer an outstanding finding. Keep the separate merge paths and #97846's stated review schedule intact. #97681 remains the umbrella; these two corrections neither implement nor authorize automatic takeover.

tui_gateway/hosted_room_replication.py
dokterdok: Composed the authored #104759 correction in d66512bfe984 (retained in f0a6cc851bd). The publisher/record owners exactly match source 2156240c0ea. The 89 focused integration checks pass, and a separate two-host check delivered queued, running and cancelled records during continuous history, including identical replay after a lost acknowledgment. No second scheduler or takeover behavior was added.

gateway/hosted_room_replica_retirement.py
dokterdok: Composed the authored d91bb120884 correction from #104601, including its installed-trigger migration and raw-SQL regressions. Both existing work-record cleanup hooks remain intact; no integration-specific guard implementation was added. The retirement and registered Disband checks pass in the 89-test integration selection.

dokterdok:
Both source review fixes are now composed and checked, including a short two-host busy-conversation test. I also carried anhtahaylove’s existing #101478 test fix with authorship intact; 103 related Desktop tests pass locally. The latest [CI attempt](https://github.com/NousResearch/hermes-agent/actions/runs/34108298346) failed before starting any jobs with a GitHub internal error. Could a maintainer rerun that workflow? The separate workflow-review label gate is unchanged; this does not request an early Desktop review.

PR #98309, opened 2026-08-30
fix(api-server): honor native persistent goals

### LINKED ISSUE #73276: [Feature Request] Support native /goal execution through /v1/runs API endpoint
### Problem or Use Case

I am integrating Hermes Agent into a custom web application through the `/v1/runs` API endpoint.

I want to provide users with the same goal-driven execution capability that is available through Hermes CLI/Gateway, especially for long-running tasks such as security investigation, log analysis, PCAP analysis, and report generation.

However, sending `/goal` commands through `/v1/runs` does not appear to activate the native Goal Loop behavior. The input is treated as a normal user prompt.

For example:
/goal Analyze the uploaded log files and PCAP files, determine whether there is an attack chain, and generate an HTML forensic report.

The model can understand the instruction, but the Goal state management, progress tracking, and goal completion loop are not triggered.

This makes it difficult for external applications to expose Hermes' goal-driven agent capabilities.

### Proposed Solution

Expose the native Goal execution flow through the API layer.

Possible approaches:

1. Make `/v1/runs` route slash commands through the same command dispatcher used by Gateway/CLI.

or

2. Add an explicit API parameter, for example:

```json
{
  "mode": "goal",
  "goal": "Analyze uploaded files and generate an attack trace report"
}

The API should preserve the native Goal behavior, including:

goal state tracking
task progress evaluation
continuation until completion
goal status query/resume support

This would allow custom Web UIs and external applications to integrate Hermes Agent without reimplementing the Goal orchestration layer.


---

# 4. Alternatives Considered

```
Currently, the alternative is to implement a custom goal orchestrator outside Hermes:

- create and store goal state
- call `/v1/runs` multiple times
- evaluate whether the goal is completed
- manage continuation and retries

However, this duplicates the native Goal Loop logic already implemented in Hermes Agent.

Exposing the existing Goal execution flow through the API would provide a more consistent developer experience.

### Alternatives Considered

_No response_

### Feature Type

Gateway / messaging improvement

### Scope

### Contribution

### Debug Report (optional)

```shell

```

### LINKED ISSUE #98299: [Bug]: Persistent /goal is documented for API server but /v1/runs bypasses GoalManager
## Summary

The persistent-goals guide says `/goal` works on the API server, including the native `wait`/park behavior for background processes. However, the structured `POST /v1/runs` path bypasses both slash-command dispatch and post-turn goal orchestration.

At `v2026.8.27` (`5fc308a70719a83cccdbba4c0e39c23f5a8239d5`), `APIServerAdapter._handle_runs()` calls `agent.run_conversation()` directly. It never constructs a `GoalManager`, dispatches `/goal`, or invokes the GatewayRunner post-turn goal continuation seam. Sending `/goal <objective>` as `input` therefore reaches the model as ordinary text and creates no `goal:<session_id>` state.

## Reproduction

1. Start `hermes gateway run` with the API server enabled.
2. Submit a structured run:

```bash
curl -X POST http://127.0.0.1:8642/v1/runs \
  -H "Authorization: Bearer $API_SERVER_KEY" \
  -H "Content-Type: application/json" \
  -d '{"session_id":"goal-api-repro","input":"/goal Wait for the background build and report when it succeeds"}'
```

3. Consume the run events and inspect `SessionDB.state_meta` / submit another turn.

Observed: one ordinary agent turn; no persistent goal state, judge verdict, continuation, or parked wait.

Expected: API-server sessions get the same native GoalManager lifecycle documented for other gateway surfaces. In particular, a background process should yield the upstream `wait` verdict, park without judge/continuation/turn consumption, and wake through the existing API-server self-post path.

## Architectural constraint

Please avoid implementing a second API-only goal loop. The current upstream implementation already owns persistence, quality gates, process/session wait barriers, timed waits, wake delivery, pause/resume/clear, and budget accounting in `GoalManager` + `GatewayRunner`. The API surface should enter that shared orchestration path or extract a shared surface-neutral coordinator.

## Suggested acceptance tests

- `/goal <objective>` through API server creates `goal:<session_id>` state and starts exactly one first turn.
- `continue` schedules the canonical continuation in the same raw API session.
- `wait_on_session`, `wait_on_pid`, and `wait_for_seconds` park with no additional model/judge call or turn charge.
- process/watch completion wakes the same raw `X-Hermes-Session-Id` through the existing self-post route.
- `status`, `pause`, `resume`, `clear`, `wait`, and `unwait` are control-plane-only where applicable.
- `/v1/runs/{id}/events` exposes goal lifecycle transitions without keeping inference active while parked.

44pie:
## Summary

- route direct API-server agent turns through a small surface bridge backed by the existing `GoalManager`
- support `/goal` lifecycle and control commands across `/v1/runs` and the shared OpenAI-compatible agent entry path
- emit native goal lifecycle events from `/v1/runs/{id}/events`
- park `wait` verdicts without extra model/judge calls or turn charges, then resume the persisted goal when the same API session is re-entered after its barrier clears
- document the API contract

This does not introduce a second goal state machine: persistence, judging, quality gates, wait barriers, continuation prompts, and budget accounting remain owned by `hermes_cli.goals.GoalManager`.

Fixes #73276.
Fixes #98299.

## Tests

- `111 passed`: full existing goal test family plus new API goal behavior tests
- `147 passed`: API server core, runs, and new goal integration tests
- `ruff check` on changed Python files
- `git diff --check`

New regression coverage includes:

- canonical `continue -> done` orchestration
- `wait` parking with no extra turn
- same-session re-entry after a cleared wait barrier
- concurrent pause during an in-flight model turn
- stop preventing judge/continuation
- control commands avoiding model calls
- `/v1/runs` SSE goal lifecycle events

### DIFF
--- a/gateway/api_goal_runtime.py
+++ b/gateway/api_goal_runtime.py
@@ -1,0 +1,213 @@
+"""Persistent-goal orchestration for stateless API-server turns.
+
+Messaging adapters let :class:`gateway.run.GatewayRunner` enqueue another
+message after a goal verdict.  The API server has no persistent outbound
+channel or adapter FIFO, so its direct agent entry points need a small bridge:
+run canonical continuation prompts in the same request until GoalManager says
+done, paused, or parked.  All state, judging, quality gates, wait barriers, and
+budget accounting remain owned by :mod:`hermes_cli.goals`.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from typing import Any, Optional
+
+
+GoalStatusCallback = Callable[[dict[str, Any]], None]
+RunTurn = Callable[[str, Optional[list[dict[str, Any]]]], dict[str, Any]]
+StopPredicate = Callable[[], bool]
+
+
+def _goal_command(text: str) -> tuple[str, str] | None:
+    stripped = (text or "").strip()
+    lowered = stripped.lower()
+    if lowered in {"/goal", "/goal status"}:
+        return ("status", "")
+    for action in ("pause", "resume", "clear"):
+        if lowered == f"/goal {action}":
+            return (action, "")
+    if lowered == "/goal unwait":
+        return ("unwait", "")
+    if lowered.startswith("/goal wait "):
+        return ("wait", stripped[len("/goal wait ") :].strip())
+    if lowered.startswith("/goal "):
+        return ("set", stripped[len("/goal ") :].strip())
+    return None
+
+
+def run_goal_aware_turn(
+    *,
+    session_id: str,
+    user_message: str,
+    conversation_history: Optional[list[dict[str, Any]]],
+    run_turn: RunTurn,
+    default_max_turns: int | None = None,
+    status_callback: GoalStatusCallback | None = None,
+    should_stop: StopPredicate | None = None,
+) -> dict[str, Any]:
+    """Run one API turn plus any native persistent-goal continuations.
+
+    Control-plane commands return without calling ``run_turn``. A judge
+    ``wait`` verdict returns immediately with the persisted barrier intact.
+    A later request for the same raw session re-enters this function; when the
+    process/session/time barrier has cleared, that request is evaluated against
+    the same persisted goal.
+    """
+    from hermes_cli.goals import GoalManager, gather_background_processes
+
+    if default_max_turns is None:
+        try:
+            from hermes_cli.config import load_config
+
+            goals_config = (load_config() or {}).get("goals") or {}
+            default_max_turns = int(goals_config.get("max_turns", 20) or 20)
+        except Exception:
+            default_max_turns = 20
+
+    manager = GoalManager(
+        session_id=session_id,
+        default_max_turns=default_max_turns,
+    )
+
+    def emit(payload: dict[str, Any]) -> None:
+        if status_callback is not None:
+            status_callback(payload)
+
+    command = _goal_command(user_message)
+    current_message = user_message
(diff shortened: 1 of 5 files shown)


Enough1122:

**Overall:** Bridges API-server stateless turns to native persistent GoalManager with continuation loop.

**What it does**
- New `gateway/api_goal_runtime.py:run_goal_aware_turn` handles `/goal` control plane (`status/pause/resume/clear/wait/unwait/set`), runs `run_turn` callback, then loops canonical continuations while `GoalManager.is_active()`, re-loading persisted state each iteration to honor concurrent pause/clear/replace and checking `should_stop`.
- Wiring in `gateway/platforms/api_server.py` for both sync and async paths: wraps `agent.run_conversation` with goal-aware runner, collects `goal_events` via `status_callback`, emits `goal.started`/`goal.status` on SSE stream, respects stop predicates.
- Adds docs and thorough tests for continuations, wait parking, re-entry after barrier, pause during flight, and control commands not calling model.

**Non-blocking notes**
- `default_max_turns` lazy-loads config with broad `except Exception` fallback to 20 — safe but swallows config parse errors silently; consider logging at debug.
- `run_goal_aware_turn` re-instantiates GoalManager each loop turn to see concurrent control changes — correct, but creates new DB handle each iteration; ensure `_DB_CACHE` reuse avoids fd churn (it does).
- Wait barrier relies on persisted `waiting_until`; tests clear it manually — document that external barrier is time/process, not automatic prod unblock.

No blocking issues found.

Issue #98298, opened 2026-08-30
[Bug]:  Heartbeat dies on every gateway restart


kclancy668-stack:
### Bug Description

/heartbeat saves state to SessionDB.state_meta but the _heartbeat_watch registry is empty on gateway boot. Active heartbeats become orphaned — they exist in the DB but never fire. Must manually /heartbeat resume after every restart to re-register.

Bug 2: Telegram adapter crashes gateway since August update

21 polling restarts since Aug 11, 5 unclean deaths. "Polling heartbeat probe failed" → reconnect timeout → updater.stop() hangs → force kill. Telegram reconnect logic destroys the gateway process instead of recovering gracefully.

Impact: Heartbeat feature is broken for 24/7 gateway users. Bot appears online but scheduled prompts never fire.

Fix needed:

Scan state_meta for active heartbeats on gateway startup and re-register watches
Harden Telegram adapter reconnect to avoid unclean gateway deaths

### Steps to Reproduce

1. set /heartbeat in an active Hermes session in either CLI/Desktop/Telegram 
2. random unexplained gateway crashes since last update
3. /heartbeat doesn't fire

### Expected Behavior

/heartbeat should fire? I'm new at this so I've never seen a successful execution to compare it to.

### Actual Behavior

The gateway keeps crashing. I'm on Windows and before the last update I had an auto-start gateway that ran 24/7 with no issues 

### Affected Component

CLI (interactive chat), Gateway (Telegram/Discord/Slack/WhatsApp)

### Messaging Platform (if gateway-related)

Telegram

### Debug Report

```shell
Debug report uploaded:
  Report       https://paste.rs/WWX3F
  agent.log    https://dpaste.com/HBE583FMG
  gateway.log  https://dpaste.com/HWVFHZ3Z5
  gui.log      https://paste.rs/WX3X6
  desktop.log  https://dpaste.com/7FW45G73Z
```

### Operating System

Windows 11

### Additional Logs / Traceback (optional)

```shell

```

### Are you willing to submit a PR for this?

- [ ] I'd like to fix this myself and submit a PR

kclancy668-stack:
[...]

PR #98024, opened 2026-08-29
fix(update): clear the fleet-restart-pending marker before the restart runs

liuhao1024:
## What does this PR do?

Fixes an infinite warning/restart loop in `hermes update` when the update is invoked from inside the gateway's own process tree (agent tool calls, webhooks, cron jobs).

`_apply_pending_fleet_restart_catchup()` cleared the `fleet_restart_pending` marker **after** `_run_pending_fleet_restart()` returned. But when the updater itself runs inside a gateway-owned process tree, the fleet restart (systemd unit restart / launchd kickstart / `kill_gateway_processes`) SIGTERMs the updater before that return, so the clear is unreachable. The marker survives forever, and every subsequent boot/update detects it, re-runs the same restart, and dies the same way — exactly the loop reported in #98010.

The fix clears the marker **before** the restart begins, and restores it only when the restart reports incomplete. Clearing early loses no safety net:

- An incomplete restart restores the marker before `sys.exit(1)`, so the next update still retries.
- The receipt-based `_receipt_reports_stale_runtime()` path in `_pending_fleet_restart_needed()` still re-triggers the catch-up if a gateway genuinely keeps serving pre-update code.

## Related Issue

Fixes #98010

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `hermes_cli/update_cmd.py` — in `_apply_pending_fleet_restart_catchup()`, move `_clear_fleet_restart_pending_marker()` to before `_run_pending_fleet_restart()`, and re-write the marker on the incomplete-restart path before exiting 1.
- `tests/hermes_cli/test_update_fleet_restart_pending.py` — add `test_catchup_clears_marker_before_restart_begins` (asserts the marker is already gone at the moment the restart executes, locking the clear-before-restart ordering that the self-kill window requires) and `test_catchup_restores_marker_when_restart_incomplete` (asserts the marker is restored and exit code 1 on an incomplete restart).

## How to Test

1. `python -m pytest tests/hermes_cli/test_update_fleet_restart_pending.py -q`
   - Observed result: `14 passed, 1 failed` — the single failure (`test_marker_written_after_pull_cleared_after_successful_restart`) also fails on a clean `upstream/main` checkout in this environment ("Managed uv install appeared to succeed but binary not found", a local uv-install issue unrelated to this diff). Both new tests pass, and all pre-existing marker/receipt tests pass.
[...]

PR #94790, opened 2026-08-25
fix(gateway): keep liveness heartbeats alive from gateway startup

fangliquanflq:
## What does this PR do?

Keeps the gateway loop-liveness heartbeat active from the beginning of gateway startup and replaces it if the task later crashes. Without this fix, startup work can fail or wedge before the heartbeat task exists, and an unexpected heartbeat-task exception silently removes the external liveness signal for the rest of the process lifetime.

### Symptom

`state/gateway.heartbeat` can retain a timestamp from a previous gateway life even while the current process remains partially operational. External supervisors then cannot distinguish the current event loop from a process that never established its liveness signal.

### Impact

Operators can receive stale health information for a live gateway process. The reported incident observed a seven-hour gateway life with no heartbeat writes; the broader blast radius is unknown.

### Bug Cause

**Trigger:** `gateway/run.py` / `GatewayRunner.start()` and `_start_loop_heartbeat_task()`

**Causal chain:**
1. Gateway startup performs diagnostics, adapter bring-up, plugin discovery, and hook setup before reaching the existing heartbeat start call.
2. An exception or wedge in that earlier work leaves the process without a heartbeat task. Separately, a heartbeat task that raises is only discarded from `_background_tasks`.
3. `state/gateway.heartbeat` remains stale with no warning or recovery, so external health checks report an invalid liveness state.

**Why it is wrong:** The heartbeat is the process-wide event-loop liveness witness, but its lifecycle was shorter and less supervised than the gateway lifecycle it represents.

**Working sibling / contrast:** The selector-floor timer and out-of-loop liveness watchdog are armed before adapters and have explicit shutdown cleanup. The heartbeat was started only after adapters connected and had no equivalent failure owner.

**Ruled out:** Heartbeat file-write exceptions alone do not terminate the task because `write_loop_heartbeat()` already handles write failures internally. The uncovered failure class is task startup timing and unexpected task-level death.

### Fix

- Arm the heartbeat alongside the other loop-liveness guards at the top of `GatewayRunner.start()`.
- Log unexpected task completion at warning level and restart it after a bounded delay.
- Cancel any pending restart handle before shutdown teardown so cleanup cannot recreate the task.
- Add lifecycle regressions for early startup failure, task-level failure recovery, and shutdown cancellation.

## Related Issue

Closes #94758

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `gateway/run.py` - start and supervise the loop heartbeat for the full gateway lifecycle.
[...]

PR #98310, opened 2026-08-30
fix(gateway): re-arm heartbeat watches after a gateway restart

chelsealong:
## What does this PR do?

Fixes the first of two bugs reported in #98298: an active `/heartbeat` becomes orphaned on every gateway restart. Bug 2 in that issue (Telegram polling reconnect killing the gateway process) is a separate, unrelated failure mode in the Telegram adapter's reconnect logic and is **not** addressed here — it needs its own investigation and fix.

### Root cause

`GatewayRunner._heartbeat_watch` (the in-memory map the poller reads to know which sessions have a due heartbeat) is process-memory only and starts empty on every gateway boot. It is only ever populated from `gateway/slash_commands.py::_handle_heartbeat_command`, i.e. only when a user issues `/heartbeat every ... / status / pause / resume`. `HeartbeatState` itself survives in `SessionDB.state_meta` across restarts, but nothing rebuilds the watch from that persisted state, so a heartbeat set before a restart never fires again until the user manually re-touches `/heartbeat` — exactly what the issue reports ("Must manually /heartbeat resume after every restart to re-register").

The docs promise better than this (`website/docs/user-guide/features/heartbeat.md`: *"firing resumes next time the session is driven"*), and the codebase already solved this exact class of problem for the sibling `/loop` feature: `LoopState` persists the originating `platform/chat_id/thread_id/user_id/user_name` as `route`, and `GatewayRunner._loop_wakeup_watcher` scans persisted loops at startup and rebuilds routing without any user interaction (see `gateway/slash_commands.py::_handle_loop_command`'s comment: *"captures the event's routing ... so the gateway's idle loop-wakeup watcher can inject ticks back into this chat even after a restart"*). `/heartbeat` never got the equivalent treatment.

### Fix

Mirrors the `/loop` pattern:

- `hermes_cli/heartbeat.py`: `HeartbeatState` gains a `route: Dict[str, str]` field (persisted/round-tripped through `to_json`/`from_json`); `HeartbeatManager.set()` accepts an optional `route`; new `list_active_heartbeats()` scans `SessionDB.state_meta` for every `active` heartbeat (same shape as `hermes_cli/loops.py::list_active_loops`).
- `gateway/slash_commands.py::_handle_heartbeat_command`: when setting a heartbeat, captures the event's routing (platform/chat/thread/user) the same way `_handle_loop_command` already does, and passes it to `mgr.set(..., route=route)`.
- `gateway/run.py`: new `_resume_heartbeat_watches()`, spawned once at startup via `_spawn_supervised` right after `_loop_wakeup_watcher`. Like `_loop_wakeup_watcher`, it's a persistent `while self._running` ticker (5s initial delay, then every 15s), not a one-shot scan — each tick only touches sessions not already in `_heartbeat_watch`, so it retries and self-heals if a given attempt hits a transient failure (adapter not yet connected, session_store not yet warm) instead of orphaning that heartbeat for the rest of the process's life. It scans `list_active_heartbeats()`, rebuilds a `MessageEvent` source from each heartbeat's persisted route via `_build_process_event_source`, and calls the existing `_register_heartbeat_watch` — so the existing 5-second poller picks the heartbeat back up with zero user interaction. CLI-owned heartbeats (no route) are left untouched; they're driven by their own watchdog thread and out of scope here.
- `gateway/run.py::_register_heartbeat_watch` docstring and `gateway/slash_commands.py`'s `/heartbeat` reply text both used to say firing only resumes when the user re-touches `/heartbeat` after a restart — updated to describe the new automatic resume instead of leaving stale, self-contradicting language next to the changed code.

## Related Issue

Addresses #98298, bug 1 only (heartbeat registry orphaned on gateway restart) — not a closing reference. #98298 also tracks bug 2 (Telegram polling reconnect can hang `updater.stop()` and kill the gateway process), which remains open and unfixed by this PR; merging this should not auto-close that issue.

## Type of Change

## Changes Made

- `hermes_cli/heartbeat.py` — `HeartbeatState.route`, `HeartbeatManager.set(route=...)`, `list_active_heartbeats()`
- `gateway/slash_commands.py` — `_handle_heartbeat_command` now persists the event's routing when setting a heartbeat; reply text updated to describe automatic resume instead of the now-stale "lives while the gateway runs" framing
- `gateway/run.py` — `_resume_heartbeat_watches()` startup task (persistent ticker, not one-shot) that re-arms `_heartbeat_watch` from persisted state; `_register_heartbeat_watch` docstring updated to match
- `tests/hermes_cli/test_heartbeat.py` — 4 new tests covering route round-trip and `list_active_heartbeats()`
- `tests/gateway/test_heartbeat_resume_watcher.py` — 5 new tests driving `_resume_heartbeat_watches` directly: registers a persisted heartbeat's route, ignores routeless heartbeats, self-heals after a transient per-tick failure, leaves already-registered sessions alone, and (end-to-end) confirms the rebuilt watch entry lets a due heartbeat actually fire through the real `_start_heartbeat_poller`

## How to Test

1. Prove the new tests fail without the fix:
   ```
   $ git checkout HEAD~1 -- hermes_cli/heartbeat.py && scripts/run_tests.sh tests/hermes_cli/test_heartbeat.py -q
   ImportError: cannot import name 'list_active_heartbeats' from 'hermes_cli.heartbeat'
   1 error in 0.18s
   $ git checkout HEAD -- hermes_cli/heartbeat.py
   ```
2. Full suite passes with the fix applied:
   ```
   $ scripts/run_tests.sh tests/hermes_cli/test_heartbeat.py -q
   [100.0% | 20/~20 | ✓30 | ✗ 0] ✓ tests/hermes_cli/test_heartbeat.py (30✓, 1.7s)
   === Summary: 1 files, 30 tests passed, 0 failed (100% complete) in 1.7s (8 workers) ===
   ```
3. The gateway-side fix (`_resume_heartbeat_watches` in `gateway/run.py`, plus the route-capture in `gateway/slash_commands.py::_handle_heartbeat_command`) is exercised directly by a new suite — `git checkout HEAD~1 -- gateway/run.py` reproduces `TypeError: got an unexpected keyword argument 'interval'` against these tests (the old code had no retry loop to test), confirming they pin the new behavior rather than passing vacuously:
   ```
   $ scripts/run_tests.sh tests/gateway/test_heartbeat_resume_watcher.py -q
   [100.0% | 5/~5 | ✓5 | ✗0] ✓ tests/gateway/test_heartbeat_resume_watcher.py (5✓, 0.9s)
   === Summary: 1 files, 5 tests passed, 0 failed (100% complete) in 0.9s (8 workers) ===
   ```
   This includes an end-to-end case: after `_resume_heartbeat_watches` re-arms the watch from a persisted route, the real `_start_heartbeat_poller` loop is driven and asserted to actually enqueue the due prompt — not just that `_heartbeat_watch` gets populated.
4. No regressions in the directly-related suites (`/loop` shares the route/rescan pattern; `test_scale_to_zero_watcher.py` exercises `_heartbeat_watch`/`_start_heartbeat_poller` directly):
   ```
   $ scripts/run_tests.sh tests/hermes_cli/test_loops.py tests/gateway/test_scale_to_zero_watcher.py tests/gateway/test_heartbeat_resume_watcher.py tests/hermes_cli/test_heartbeat.py -q
   === Summary: 4 files, 127 tests passed, 0 failed (100% complete) in 4.6s (8 workers) ===
   ```
5. `ruff check` on the changed files (including the new test file): `All checks passed!`

### Code

- [x] I searched for existing PRs to make sure this isn't a duplicate (checked #92656, #94790, #98024, #80185 — none touch the gateway-restart heartbeat orphaning this issue reports)
- [x] I've run the repo's test entry on relevant tests and all tests pass
- [ ] I've tested on my platform — tested on Linux only; the issue was reported on Windows and I don't have a Windows environment to verify against, but the change is platform-agnostic (no OS-specific code paths)

- [x] `website/docs/user-guide/features/heartbeat.md` already describes the intended behavior ("firing resumes next time the session is driven"); this PR makes the code match it, so no doc-file edit needed. It does update two in-code/in-product strings that pre-dated this fix and had gone stale relative to it: the `_register_heartbeat_watch` docstring in `gateway/run.py` and the `/heartbeat` command's user-facing reply in `gateway/slash_commands.py` (both previously said firing only resumes after a user re-touches `/heartbeat`, which this PR makes untrue).
- [x] Config example updates are N/A
- [x] Contributor guide updates are N/A
- [x] I've considered cross-platform impact; no OS-specific code paths touched
- [x] Tool description and schema updates are N/A

## AI Disclosure

This PR was authored with AI assistance (Claude). The root-cause analysis, fix, and tests were reviewed against the existing `/loop` implementation before writing, and verified locally (test-fails-without-fix, full relevant suite, ruff) as described above.

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -13850,8 +13850,10 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
 
         # Re-arm heartbeat watches orphaned by this restart (the in-memory
         # registry starts empty every boot; see _resume_heartbeat_watches).
-        # One-shot: it returns after the scan, so _spawn_supervised's
-        # clean-exit rule correctly never respawns it.
+        # Like _loop_wakeup_watcher above, this ticks for the life of the
+        # process rather than scanning once, so a transient startup race
+        # (adapter not yet connected, session_store not yet warm) self-heals
+        # on the next tick instead of orphaning that heartbeat permanently.
         self._spawn_supervised(self._resume_heartbeat_watches, "heartbeat_resume_watcher")
 
         # Start the scale-to-zero idle watcher ONLY when this instance is opted
@@ -22522,8 +22524,10 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
         The registry maps ``quick_key`` → ``(source, session_id)`` so the
         poller can rebuild a MessageEvent and enqueue via the adapter FIFO.
         In-memory by design: heartbeat STATE survives restarts in SessionDB,
-        but firing resumes when the user touches /heartbeat again in the new
-        gateway process (documented; durable schedules belong to cron).
+        and firing resumes automatically once the new gateway process comes
+        back up — ``_resume_heartbeat_watches`` calls this same method for
+        every persisted heartbeat at startup, so no user action is required
+        (durable schedules that must survive anything still belong to cron).
         """
         watch = getattr(self, "_heartbeat_watch", None)
         if watch is None:
@@ -22972,50 +22976,64 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                 logger.debug("loop wakeup watcher error: %s", exc)
             await asyncio.sleep(interval)
 
-    async def _resume_heartbeat_watches(self) -> None:
+    async def _resume_heartbeat_watches(self, interval: float = 15.0) -> None:
         """Re-arm ``_heartbeat_watch`` from persisted state after a restart.
 
         The watch dict is process-memory only and starts empty on every
-        gateway boot, so an active heartbeat set before a restart was
-        otherwise orphaned — its state survives in SessionDB, but nothing
+        gateway boot, so an active heartbeat set before a restart would
+        otherwise be orphaned — its state survives in SessionDB, but nothing
         polls it — until the user ran /heartbeat again. Gateway heartbeats
         persist their originating platform/chat/thread in
-        ``HeartbeatState.route`` (mirrors /loop's restart recovery), so this
-        rebuilds the watch without any user interaction. Runs once at
-        startup; a CLI-owned heartbeat carries no route and is driven by its
-        own watchdog thread instead.
+        ``HeartbeatState.route`` (mirrors /loop's restart recovery via
+        ``_loop_wakeup_watcher``), so this rebuilds the watch without any
+        user interaction.
+
+        Like ``_loop_wakeup_watcher``, this is a ``while self._running``
+        ticker rather than a one-shot scan: a transient failure on any
+        given tick (adapter not yet connected, session_store not yet warm)
+        would otherwise orphan that heartbeat for the rest of the process's
+        life, with nothing left to retry it. Each tick only touches
+        sessions not already present in ``_heartbeat_watch``, so a heartbeat
+        that's already re-armed is left alone. A CLI-owned heartbeat carries
+        no route and is driven by its own watchdog thread instead.
         """
         await asyncio.sleep(5)  # let platforms finish connecting
-        try:
-            from hermes_cli.heartbeat import list_active_heartbeats
+        while self._running:
+            try:
+                from hermes_cli.heartbeat import list_active_heartbeats
 
-            await self._warm_goals_session_db("heartbeat resume")
-            for sid, state in list_active_heartbeats():
-                route = state.route or {}
-                platform_name = route.get("platform", "")
-                chat_id = route.get("chat_id", "")
-                if not platform_name or not chat_id:
-                    continue
-                evt_stub = {
-                    "session_key": "",
-                    "platform": platform_name,
-                    "chat_id": chat_id,
-                    "chat_type": route.get("chat_type", ""),
-                    "thread_id": route.get("thread_id", ""),
-                    "user_id": route.get("user_id", ""),
-                    "user_name": route.get("user_name", ""),
-                }
-                source = self._build_process_event_source(evt_stub)
-                if source is None:
-                    continue
-                try:
-                    quick_key = self._session_key_for_source(source)
-                except Exception:
-                    continue
-                if quick_key:
-                    self._register_heartbeat_watch(quick_key, source, sid)
-        except Exception:
-            logger.debug("heartbeat resume scan failed", exc_info=True)
+                await self._warm_goals_session_db("heartbeat resume")
+                watch = getattr(self, "_heartbeat_watch", None) or {}
+                resumed_sids = {sid for _source, sid in watch.values()}
+                for sid, state in list_active_heartbeats():
+                    if sid in resumed_sids:
+                        continue
+                    route = state.route or {}
+                    platform_name = route.get("platform", "")
+                    chat_id = route.get("chat_id", "")
+                    if not platform_name or not chat_id:
+                        continue
+                    evt_stub = {
+                        "session_key": "",
+                        "platform": platform_name,
+                        "chat_id": chat_id,
+                        "chat_type": route.get("chat_type", ""),
+                        "thread_id": route.get("thread_id", ""),
+                        "user_id": route.get("user_id", ""),
+                        "user_name": route.get("user_name", ""),
+                    }
+                    source = self._build_process_event_source(evt_stub)
+                    if source is None:
+                        continue
+                    try:
+                        quick_key = self._session_key_for_source(source)
+                    except Exception:
+                        continue
+                    if quick_key:
+                        self._register_heartbeat_watch(quick_key, source, sid)
+            except Exception:
+                logger.debug("heartbeat resume scan failed", exc_info=True)
+            await asyncio.sleep(interval)
 
     @staticmethod
     def _get_guild_id(event: MessageEvent) -> Optional[int]:
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -13848,6 +13848,12 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
         # originating chats while the session is idle.
         self._spawn_supervised(self._loop_wakeup_watcher, "loop_wakeup_watcher")
 
+        # Re-arm heartbeat watches orphaned by this restart (the in-memory
+        # registry starts empty every boot; see _resume_heartbeat_watches).
+        # One-shot: it returns after the scan, so _spawn_supervised's
+        # clean-exit rule correctly never respawns it.
+        self._spawn_supervised(self._resume_heartbeat_watches, "heartbeat_resume_watcher")
+
         # Start the scale-to-zero idle watcher ONLY when this instance is opted
         # in (the NAS "Labs" HERMES_SCALE_TO_ZERO stamp), messaging is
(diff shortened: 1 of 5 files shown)


andrexibiza (commented):
Reviewed exact head `e34c6554feb0f0e8f8dcd52d33d7a6ecee0dff23` against the live issue/PR graph and current routing/session contracts. The retrying watcher is a good improvement over a one-shot startup scan, and the exact-head hosted receipts are green: CI `33289618066`, Docker `33289617632`, and Nix `33289617639` all succeeded. I do not think this object is ready to land yet, for three concrete reasons.

### P1 — existing persisted heartbeats still do not recover after upgrading to this fix

The bug report is specifically about heartbeats that already exist in `SessionDB.state_meta` before a gateway restart. This head adds `HeartbeatState.route`, but backward-compatible deserialization intentionally defaults old rows to `route={}`. `_resume_heartbeat_watches()` then does:

```py
route = state.route or {}
platform_name = route.get("platform", "")
chat_id = route.get("chat_id", "")
if not platform_name or not chat_id:
    continue
```

So every pre-existing heartbeat row — exactly the population affected by #98298 — is skipped after the user installs this change. The current `/heartbeat resume` path only calls `mgr.resume()` and re-registers the in-memory watch; it does not persist the current `event.source` into the new route field. That means even the documented workaround does not migrate the row for the *next* restart. The user would have to recreate the heartbeat with `/heartbeat every ...` before this new recovery path can ever help it.

The new test `test_state_roundtrip_defaults_route_to_empty_dict()` correctly demonstrates the legacy representation, but there is no upgrade regression proving that a legacy active row actually re-arms. We need that test and a migration/fallback path. The simplest existing source of truth is the persisted session-routing entry that already owns `session_id -> SessionSource`; see competing #98313 below.

### P1 — `HeartbeatState.route` creates a second, incomplete routing authority

`SessionEntry.origin` is already durable routing metadata and serializes the full `SessionSource`. `SessionSource` currently carries qualified routing dimensions that the new heartbeat-local dict drops, including `profile` (explicitly documented as driving session-key namespacing and per-turn config/credential scope), `scope_id`/legacy `guild_id` (server/workspace isolation), parent/alternate IDs, and other platform-specific fields.

This head persists only:

```text
platform, chat_id, chat_type, thread_id, user_id, user_name
```

then reconstructs a source from that reduced snapshot and derives a fresh `quick_key`. In a multiplexed/profile-qualified or workspace-qualified route, restart recovery can therefore rebuild a different identity than the one that originally owned the session. That is the other side of the restart-recovery shape: durable *schedule state* is not enough; the delivery identity used to re-arm it must remain the canonical qualified identity too.

Open #98313 by `liuhao1024` is competing closure work for the same bug and directly collides in `gateway/run.py`, but it is architecturally useful here because it restores from the existing persisted `SessionEntry.origin`/`entry.session_key` instead of minting another route schema. #98310 has the stronger transient-failure behavior because its supervised ticker retries; #98313 has the stronger ownership source and naturally works for legacy heartbeat rows. I would converge those properties rather than land both implementations or choose one contributor's work as disposable: canonical session-routing ownership + the self-healing retry loop, with both contributors' lineage preserved.

Required regressions should include at least: (1) an active legacy heartbeat row with no `route` survives restart; (2) a profile-qualified/multiplexed session re-arms the exact persisted session key and source; (3) a transient first scan failure retries and eventually restores; and (4) paused/cleared/CLI-only state does not get spuriously armed.

### P2 — the closing reference would close an issue whose second bug remains live

#98298 contains two independent failures: the heartbeat registry orphaning and a Telegram polling/reconnect path that can hang `updater.stop()` and kill the gateway. This PR correctly says in prose that bug 2 is out of scope, but `Fixes #98298 (bug 1 only ...)` is still a GitHub closing reference for the *whole issue*. Merging it would make the tracker say #98298 is resolved while the Telegram crash remains unfixed.

Please either split the two defect classes and close only the heartbeat-specific issue, or change this PR to a non-closing reference and link the follow-up owner for bug 2. The original reporter's two observations should stay attributable; they just need distinct closure objects.

### Landing topology / exact-object note

This head and #98313 are not complementary merge-order layers as written; they are competing implementations of the same restart-rearm owner and both modify `gateway/run.py`. One should become the convergence carrier, explicitly crediting the other's unique mechanism. Also, the PR was cut from base `52e5e7c034edfbd0105d0b9c315c4eecb6be7f5e`, while live `main` has advanced to `58523f284ca52a162a213a7efd335b203e783706`. After convergence, recompose on current main and require fresh exact-head CI/Docker/Nix receipts; the current green runs prove this exact `e34c6554...` object, not the eventual landing object.

There is strong work here: the root-cause diagnosis is correct, the retrying supervised watcher is the right liveness instinct, and the end-to-end poller regression is much better than merely asserting that a dict entry appears. The remaining work is to make the recovery apply to the users who already have broken heartbeats, and to restore through one qualified routing authority rather than a second partial snapshot. Once those are closed, this can be a durable fix rather than just a forward-only one. 🚀

chelsealong:
Pushed 6d9495bac addressing the two P1s: `_resume_heartbeat_watches` now resolves routing via `_gateway_session_origin_for_id` (the same durable `SessionEntry.origin`/`SessionSource` lookup `/resume` already uses) instead of the heartbeat-local `route` dict, which is now removed entirely from `HeartbeatState`/`HeartbeatManager`/`_handle_heartbeat_command`. This means pre-existing heartbeat rows (no route ever persisted) recover too, and the source used to re-arm is the session's actual qualified `SessionSource` rather than a reconstruction that dropped `profile`/`scope_id`. Added regression tests for a legacy no-route row, for a no-origin (CLI-only) session being left alone, and that the origin's source object is passed through unmodified rather than rebuilt (`tests/gateway/test_heartbeat_resume_watcher.py`, `tests/hermes_cli/test_heartbeat.py`); full relevant suite is green (129 passed) and `ruff check` is clean on changed files.

Also updated the PR body so `#98298` is a non-closing reference, since bug 2 (Telegram reconnect) is still open.

Left alone: converging with #98313 and rebasing onto current `main` — that's a maintainer call on which implementation becomes the carrier, not something to resolve unilaterally from this side.

liyangbing:
The persisted route plus startup re-arm mirrors the right durable-state pattern. A few invariants could keep restart recovery safe under concurrent ticks and configuration changes:\n\n- Bind each heartbeat watch to a session generation and a stable heartbeat id. A late tick from the pre-restart watcher must not enqueue into a replaced session or duplicate a wake already accepted by the new watcher.\n- Treat persisted routing as delivery authority, not merely display metadata. Revalidate that the platform/chat/thread target is still authorized and available before enqueueing; if it is not, record a bounded deferred reason instead of retrying indefinitely.\n- Make the scan and registration idempotent across overlapping startup tasks, transient store failures, and multiple gateway instances. Assert at most one active watcher per heartbeat/session tuple and one due tick per interval window.\n- Test clock jumps, missed intervals, disabled or expired heartbeats, malformed legacy route data, and a heartbeat edited while the resume scan is in flight. Persist the observed revision so stale recovery cannot resurrect an old schedule or route.\n- Keep the first post-restart wake auditable: heartbeat id, session id, route revision, scheduler generation, due time, and enqueue result. Do not log message content or credentials.\n- Add a shutdown/restart crash matrix around route persistence and watch registration, and verify no orphaned task remains when the gateway stops.\n\nThe handbook's durable-session guidance covers restart recovery, generation fencing, and bounded wakeups: https://sandbaseai.github.io/deepseek-harness-handbook/session-log-durability.html

PR #98314, opened 2026-08-30
fix(tui): restore DB-backed history for Desktop sessions

wang0090:
## Summary

Fix Desktop/TUI multi-turn history loss caused by hollow in-memory session history.

The persisted session DB contains the full prior user/assistant content, but the Desktop/TUI turn path feeds `run_conversation` from the gateway's in-memory `session["history"]`. In affected sessions, those entries retain the correct message count while their prior content is hollow, so provider input tokens remain flat across turns and the model behaves as if no earlier context exists.

The CLI path does not reproduce this because it loads authoritative DB-backed history via `SessionDB.get_messages_as_conversation`.

This change reconciles hollow in-memory history with the persisted DB transcript before the Desktop/TUI turn is submitted.

## Root Cause

Observed before the fix on real Desktop sessions:

- history count: `0 → 2 → 4`
- provider prompt tokens: `2050 → 2050 → 2050`
- persisted DB messages: present with real content

So the session and persistence layers were intact, but prior message content was not reaching the provider.

The Desktop/TUI submit path relied on in-memory `session["history"]`, while the authoritative DB transcript remained complete.

## Changes

- add a history reconciliation helper in `tui_gateway/server.py`
- detect hollow prior entries before Desktop/TUI prompt submission
- recover real prior content from `SessionDB.get_messages_as_conversation`
- preserve any unsaved in-memory tail
- leave healthy in-memory history unchanged
- add focused regression coverage for reconciliation behavior

## Validation

Static:
- `py_compile`: PASS
- `git diff --check`: PASS

Focused regression:
- intact history remains unchanged: PASS
- hollow history is restored from DB: PASS
- unsaved tail is preserved: PASS
- empty history handling: PASS
- DB unavailable/fallback behavior: PASS

Desktop-turn integration:
- hollow `session["history"]` simulated
- `run_conversation` receives real DB-backed prior content: PASS

Real Desktop 3-turn validation (`qwen3.5:9b`):

Before fix:
- prompt tokens: `2050 → 2050 → 2050`

After fix:
- Round 1: `history=0`, `in=16427`
- Round 2: `history=6`, `in=16835`
- Round 3: `history=8`, `in=16898`

Semantic continuity also passed:
- Round 2 correctly recalled all prior values
- Round 3 retained both newly supplied information and earlier conversation context

The larger history counts in the real test include tool-call/tool-result messages and are expected.

## Scope

This fix is limited to the Desktop/TUI gateway history handoff.

It does not change:
- CLI history loading
- Ollama configuration
- session DB schema
- model configuration
- provider routing

This PR is intentionally separate from the local Ollama `num_ctx` transport fix in PR #98229 because the two issues have different root causes.

### DIFF
--- a/tests/tui_gateway/test_history_reconcile.py
+++ b/tests/tui_gateway/test_history_reconcile.py
@@ -1,0 +1,137 @@
+"""Regression test for the Desktop/TUI history-hollow fix.
+
+Root cause (HISTORY_PIPELINE_BUG_CONFIRMED): the Desktop/TUI gateway feeds
+``run_conversation`` an in-memory ``session["history"]`` that can be hollowed
+across turns (prior-turn content arrives empty while the session DB still holds
+the real text). The CLI surface loads history from
+``SessionDB.get_messages_as_conversation`` every turn and always carries real
+content. ``_reconcile_history_with_session_db`` makes the Desktop path do the
+same: when the in-memory copy is hollow, fall back to the DB-backed transcript.
+
+These tests exercise the REAL helper (and the exact history-handling sequence
+inside ``_run_prompt_submit``) with a fake DB/agent so they run offline and fast.
+"""
+import threading
+import types
+
+import pytest
+
+from tui_gateway import server as S
+
+
+class FakeDB:
+    def __init__(self, rows):
+        self._rows = rows
+
+    def get_messages_as_conversation(self, session_id, include_ancestors=False,
+                                      repair_alternation=False, include_row_ids=False):
+        return [dict(r) for r in self._rows]
+
+
+REAL = [
+    {"role": "user", "content": "remember the code is 4271"},
+    {"role": "assistant", "content": "Got it, code 4271 noted."},
+]
+
+
+def _agent(db):
+    a = types.SimpleNamespace()
+    a._session_db = db
+    a.session_id = "sess_key"
+    return a
+
+
+def _session():
+    return {"session_key": "sess_key", "history": [], "history_lock": threading.Lock()}
+
+
+def test_intact_history_returned_unchanged():
+    agent = _agent(FakeDB(REAL))
+    sess = _session()
+    hist = [dict(r) for r in REAL]
+    assert S._reconcile_history_with_session_db(agent, sess, hist) == REAL
+
+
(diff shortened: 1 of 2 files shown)


liyangbing:
<h3>Review</h3>
<p>The evidence cleanly separates an intact persisted transcript from a hollow in-memory handoff, and the focused reconciliation scope is appropriate. The important invariant is that DB recovery must not overwrite newer unsaved work or turn a stale snapshot into the new authority.</p>
<h3>Suggested contract</h3>
<ul>
<li>Bind reconciliation to the Session identity, profile, and a history generation or revision; reject a DB snapshot that is older than the in-memory durable frontier.</li>
<li>Define the merge boundary explicitly: recover hollow persisted entries, preserve an unsaved tail in order, and avoid duplicating entries when a prior submit partially committed before the retry.</li>
<li>Make the DB read and in-memory replacement atomic from the turn submitter's perspective, or keep the old history until the recovered snapshot passes shape, ordering, and identity checks.</li>
<li>Keep provider input projection separate from canonical Session storage so token counts and model-facing content can be compared without treating a rendered prompt as durable truth.</li>
<li>Log redacted recovery evidence—Session identity, old/new revision, recovered count, tail count, and decision—without message text, credentials, or workspace paths.</li>
</ul>
<h3>Regression matrix</h3>
<ul>
<li>healthy history, hollow history, empty history, mixed hollow entries, and an unsaved in-memory tail</li>
<li>DB unavailable, corrupt or stale transcript, moved profile, Session switch during submit, and concurrent turn admission</li>
<li>partial append before crash, duplicate retry, out-of-order rows, multimodal content, tool-call/result pairs, and exact ordering preservation</li>
<li>Desktop, TUI, and CLI paths followed by restart and replay, verifying prompt growth, semantic continuity, and no duplicate context</li>
</ul>
<p>The handbook's <a href="https://sandbaseai.github.io/deepseek-harness-handbook/session-history-recovery.html">Session-history recovery guide</a> and <a href="https://sandbaseai.github.io/deepseek-harness-handbook/session-log-durability.html">Session-log durability guide</a> cover the same authority, merge, revision, and restart boundaries.</p>

PR #98315, opened 2026-08-30
fix(agent): uppercase finish_reason (STOP/MAX_TOKENS) no longer bypasses stop/length handling (port of omp#9566)

teknium1:
## Summary
Uppercase finish reasons (`STOP`, `MAX_TOKENS`) from OpenAI-compatible gateways fronting Gemini backends no longer bypass Hermes' stop/length handling — they are folded to the OpenAI contract at wire intake.

Port of [can1357/oh-my-pi#9566](https://github.com/can1357/oh-my-pi/pull/9566) (@yiliang114). Root cause: every downstream comparison in Hermes uses lowercase literals (`== "stop"`, `== "length"`, `== "content_filter"`, `== "tool_calls"`), so an uppercase reason silently fell through — a clean `STOP` completion missed stop handling and a `MAX_TOKENS` truncation never entered the length-recovery path.

## Changes
- `agent/message_sanitization.py`: `normalize_finish_reason()` — single owner; case fold + alias map (`max_tokens`→`length`, `end`→`stop`, `function_call`→`tool_calls`). Non-string/empty values pass through unchanged.
- `agent/transports/chat_completions.py`: fold at `normalize_response()` (non-streaming intake); Poolside int-reason path preserved.
- `agent/chat_completion_helpers.py`: fold at the streaming chunk-capture site.
- `tests/agent/transports/test_finish_reason_normalization.py`: 23 tests — owner table, both choke points, alias-identity check, pass-through invariants.

## Adaptation vs upstream
omp normalizes inside its `mapStopReason` switch in the openai-completions provider. Hermes' comparisons are spread across `conversation_loop.py` / `chat_completion_helpers.py`, so instead of patching a switch we add a single normalizer at the two wire-intake choke points — every downstream comparison site is covered without touching any of them (never-patch-predicates: fix the class at the choke point).

## Validation
| | Before | After |
|---|---|---|
| `normalize_response(finish_reason="STOP")` | `"STOP"` (falls through all handling) | `"stop"` |
| `normalize_response(finish_reason="MAX_TOKENS")` | `"MAX_TOKENS"` (no length recovery) | `"length"` |
| lowercase contract values | unchanged | byte-identical pass-through |
| Sabotage run (fix reverted) | — | 2 tests fail as expected |
| `tests/agent/transports/` targeted subset | — | 80 passed |
| `tests/agent/test_*sanitiz*` | — | 56 passed |

## Infographic

finish reason normalization


liyangbing:
<h3>Review</h3>
<p>Normalizing gateway-specific finish reasons at the wire-intake choke points is the right way to protect the downstream stop and length contracts. The critical invariant is that streaming and non-streaming paths produce the same canonical reason without erasing unknown provider information.</p>
<h3>Suggested contract</h3>
<ul>
<li>Define the canonical mapping table once and apply it before any stop, length-recovery, retry, or tool-call decision; keep the original wire value available in redacted diagnostics.</li>
<li>Make non-streaming responses and streamed terminal chunks converge through the same normalizer, including mixed-case values, aliases, missing values, and non-string provider extensions.</li>
<li>Keep <code>stop</code>, <code>length</code>, <code>content_filter</code>, and <code>tool_calls</code> semantically distinct; an unknown reason should remain observable and must not silently become a successful stop.</li>
<li>Ensure normalization is idempotent and does not mutate shared response objects or alter the Poolside integer path; downstream consumers should see one stable contract.</li>
<li>When <code>length</code> triggers recovery or compaction, preserve the incomplete-attempt evidence and prevent a partial summary or duplicate retry from being committed as a completed turn.</li>
</ul>
<h3>Regression matrix</h3>
<ul>
<li>uppercase, lowercase, mixed-case, alias, empty, missing, non-string, and unknown finish reasons</li>
<li>streaming terminal chunk, non-streaming response, provider retry, tool-call termination, content filtering, and Poolside integer reason</li>
<li>normalization called twice, shared-object reuse, concurrent streams, and a response carrying extra provider metadata</li>
<li><code>length</code> followed by recovery or compaction, interrupted retry, failed summary, successful retry, and exact terminal ledger evidence</li>
</ul>
<p>The handbook's <a href="https://sandbaseai.github.io/deepseek-harness-handbook/output-token-limit-reached.html">output-limit guide</a> and <a href="https://sandbaseai.github.io/deepseek-harness-handbook/compaction-summary-truncated.html">compaction-boundary guide</a> document the same canonical finish-reason, incomplete-checkpoint, and retry boundaries.</p>

PR #98316, opened 2026-08-30
fix(desktop): isolate session scroll and composer focus

mor44-AI:
## Summary
- scope jump-to-bottom requests to the originating session instead of broadcasting to every mounted transcript
- prevent hidden kept-alive panes from stealing composer focus on remount or runtime rebind
- add regressions for cross-session scroll routing and hidden-pane autofocus

## Verification
- npm run typecheck --workspace apps/desktop
- targeted Vitest: 26 tests passed

### DIFF
--- a/apps/desktop/src/app/chat/composer/hooks/use-composer-draft.test.tsx
+++ b/apps/desktop/src/app/chat/composer/hooks/use-composer-draft.test.tsx
@@ -2,6 +2,7 @@ import { act, cleanup, render, waitFor } from '@testing-library/react'
 import { useLayoutEffect } from 'react'
 import { afterEach, describe, expect, it, vi } from 'vitest'
 
+import { PaneVisibleContext } from '@/components/pane-shell/pane-visibility'
 import { clearSessionDraft, type ComposerAttachment, mainComposerScope, stashSessionDraft } from '@/store/composer'
 import { $connection } from '@/store/session'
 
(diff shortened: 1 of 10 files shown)


liyangbing:
<h3>Review: verify visibility state is scoped as well as scroll requests</h3><p>The new <code>requestScrollToBottom</code> handler map is keyed by <code>sessionId</code>, which prevents the click action from broadcasting to every mounted transcript. However, <code>$threadScrolledUp</code> and <code>$threadJumpButtonVisible</code> in <code>apps/desktop/src/store/thread-scroll.ts</code> remain singleton atoms, while this PR explicitly keeps hidden panes mounted and adds session-swap/remount coverage.</p><p>Please make the ownership invariant explicit in a test: mount two sessions, leave A scrolled up, keep B hidden, then change B's stick-to-bottom state and assert that the visible A composer/status stack/jump button still reflects A. Repeat with the active session changing while both panes remain mounted. If only the active pane is allowed to publish the atoms, enforce that at the publishing boundary; if multiple panes can publish, make the visibility state session-keyed and have consumers select the active key.</p><h3>Related lifecycle checks</h3><p>Also cover an unmount/remount sequence where the old session's cleanup runs after the new session has registered, and assert that cleanup cannot clear the new session's handler or visibility state. The existing “routes a scroll request only to its session” test proves action routing, but not the state-rendering path that determines whether the control is visible or incorrectly says “Approval needed”.</p><p>This is a narrow follow-up to the PR's stated goal: request routing and visual state need the same session ownership rule, especially with hidden keep-alive panes.</p>

OutThisLife:
Superseded by #99881 (this one was already closed). The session-keyed jump routing landed there, plus the visibility-gated atoms the review here asked for. Credit carried via `Co-authored-by`.

PR #98272, opened 2026-08-30
feat: /plan survives the platform command-menu caps as a built-in command (closes #67264, #36821, salvage #67292)

teknium1:
## Summary

`/plan` now shows up in every platform's command menu because it is a built-in command, not a skill. The bundled `plan` skill's auto-generated slash command was the poster child for the capped-menu bug: skills are the only tier trimmed at the Telegram (60-slot default) and Discord (100 hard cap) menus, alphabetically — `plan` sat at index 57 of 82 bundled skills and silently fell off for most installs. Core `CommandDef` commands are never trimmed.

Salvages #67292 by @webtecnica (first `/plan` submission, issue #67264), reworked from an inline planning prompt to the `/learn`-style prompt-injection pattern with workspace-saved plans. Closes #67264. Closes #36821 (empty `/plan` infers the task from conversation context).

## Changes

- `agent/plan_prompt.py` (new): `build_plan_prompt()` — plan-mode ground rules + plan-authoring craft distilled from the retired skill. No engine, no model-tool footprint, cache-safe (no system-prompt or history mutation), works on any terminal backend.
- `hermes_cli/commands.py`: `CommandDef("plan", ..., args_hint="[task]")` — gateway-visible, replaces the contributor's `cli_only` version.
- `cli.py` + `hermes_cli/cli_commands_mixin.py`: `_handle_plan_command` injects the prompt onto `_pending_input` (same as `/learn`).
- `gateway/run.py`: `/plan` branch sends an ack, rewrites `event.text`, and falls through — role alternation preserved.
- `tui_gateway/methods_tools.py`: `command.dispatch` branch (`"plan"` was already in `_PENDING_INPUT_COMMANDS`).
- Removed `skills/software-development/plan/`; `PROTECTED_BUILTIN_SKILLS` is now empty (mechanism kept; dependent curator/usage tests use monkeypatched sentinels).
- Docs: slash-commands rows (EN + zh-Hans), skills-catalog rows, sidebar entry, generated skill pages deleted, `related_skills` references cleaned, curator/skills/tips prose updated.

## Validation

| | Before (origin/main) | After |
|---|---|---|
| `resolve_command("plan")` | `None` (skill only) | `CommandDef(plan)` |
| Telegram menu (default 60-cap) | `plan` absent (trimmed) | `plan` present, core tier |
| CLI `/plan <task>` | loads skill if visible | prompt injected via `_pending_input` (E2E) |
[...]

PR #98317, opened 2026-08-30
feat(telegram): every command and skill is searchable via @botname — inline picker bypasses the menu cap

teknium1:
## Summary

Typing `@yourbotname <query>` in any Telegram chat now opens a live, searchable picker over **every** Hermes command and installed skill — the command-menu cap stops mattering. The BotCommand menu is hard-capped by Telegram (100/scope, ~4KB payload; Hermes defaults to 60 slots) so most skill commands can never appear in the `/` menu; inline mode computes results per keystroke and paginates 50 at a time, exactly the trick Discord's dynamic `/skill` autocomplete uses (#18741). This is the third leg of the menu-cap work after #98272 (/plan → built-in) and #98280 (priority before cap).

Tapping a result sends the command text (`/plan migrate auth`) into the chat as the user; command-prefixed messages reach the bot even under default privacy mode, so dispatch flows through the existing command path — zero new dispatch code.

## Changes

- `plugins/platforms/telegram/inline_picker.py` (new): PTB-free catalog/rank/pagination logic. Catalog = core gateway commands + plugin commands + ALL skills via the shared collector with `max_slots=None` (nothing trimmed). First query token filters (prefix > name-substring > description, hyphen/underscore-equivalent); the remainder rides into the sent command as its argument.
- `plugins/platforms/telegram/adapter.py`: `InlineQueryHandler` registration + `_handle_inline_query`. Auth reuses `_is_callback_user_authorized` (allowlists/pairing/multiplex); unauthorized users get an empty list so the installed-skill catalog is not leaked (inline queries arrive from ANY chat, including ones the bot is not in). `is_personal=True` so Telegram never shares cached pages across users.
- Docs: `telegram.md` inline-picker section, incl. the one-time BotFather `/setinline` setup. The handler is inert until inline mode is enabled — Telegram never delivers `inline_query` updates otherwise — so registering unconditionally is safe.

## Validation

| | Result |
|---|---|
| E2E catalog (temp HERMES_HOME, 72 skills) | 138 entries — vs 60-slot menu; `zzz-last-skill` present |
| `@bot songsee analyze this track` | tap sends `/songsee analyze this track` |
| Pagination walk | all pages sum to full catalog, `next_offset=""` terminates |
| Real adapter handler (conftest telegram mock) | answers with articles, `is_personal=True` |
| Unauthorized / missing user | empty result list |
| Tests | 31 passed (`test_telegram_inline_picker` ×19 new, `test_telegram_conflict`, `test_telegram_forum_commands`); ruff clean |

**Live repro:** on main there is no inline handler at all (`InlineQueryHandler` absent from the adapter) — `@botname` queries go unanswered; on this branch the full E2E above returns the complete 138-entry catalog with search, arg carry-through, and pagination.

## Infographic

telegram inline picker — no more menu cap

### DIFF
--- a/plugins/platforms/telegram/adapter.py
+++ b/plugins/platforms/telegram/adapter.py
@@ -150,6 +150,7 @@ try:
         Application,
         CommandHandler,
         CallbackQueryHandler,
+        InlineQueryHandler,
         MessageHandler as TelegramMessageHandler,
         ContextTypes,
         TypeHandler,
@@ -169,6 +170,7 @@ except ImportError:
     Application = Any
     CommandHandler = Any
     CallbackQueryHandler = Any
+    InlineQueryHandler = Any
     TypeHandler = Any
     TelegramMessageHandler = Any
     HTTPXRequest = Any
@@ -342,7 +344,7 @@ def check_telegram_requirements() -> bool:
     """
     global TELEGRAM_AVAILABLE, Update, Bot, Message, InlineKeyboardButton
     global InlineKeyboardMarkup, LinkPreviewOptions, Application
-    global CommandHandler, CallbackQueryHandler, TelegramMessageHandler
+    global CommandHandler, CallbackQueryHandler, InlineQueryHandler, TelegramMessageHandler
     global ContextTypes, filters, ParseMode, ChatType, HTTPXRequest, TypeHandler
     if TELEGRAM_AVAILABLE:
         return True
@@ -361,6 +363,7 @@ def check_telegram_requirements() -> bool:
         from telegram.ext import (
             Application as _App, CommandHandler as _CH,
             CallbackQueryHandler as _CQH,
+            InlineQueryHandler as _IQH,
             MessageHandler as _MH,
             ContextTypes as _CT, filters as _filters,
             TypeHandler as _TH,
@@ -378,6 +381,7 @@ def check_telegram_requirements() -> bool:
     Application = _App
     CommandHandler = _CH
     CallbackQueryHandler = _CQH
+    InlineQueryHandler = _IQH
     TelegramMessageHandler = _MH
     ContextTypes = _CT
     filters = _filters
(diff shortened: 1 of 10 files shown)


liyangbing:
<h3>Review</h3>

<p>The feature is well-scoped in head <code>778f3d40e656bd5ae72f1beef5dae8f51a1771e2</code>: the picker is uncapped and PTB-free in <code>inline_picker.py</code>, while dispatch still uses the normal command path.</p>

<ul>
<li><a href='https://github.com/NousResearch/hermes-agent/blob/778f3d40e656bd5ae72f1beef5dae8f51a1771e2/plugins/platforms/telegram/adapter.py#L4381-L4389'>The handler is registered unconditionally but remains inert until BotFather inline mode is enabled.</a></li>
<li><a href='https://github.com/NousResearch/hermes-agent/blob/778f3d40e656bd5ae72f1beef5dae8f51a1771e2/plugins/platforms/telegram/adapter.py#L7209-L7301'>Authorization deliberately runs without inline chat context and answers unauthorized users with an empty personal result set.</a></li>
<li><a href='https://github.com/NousResearch/hermes-agent/blob/778f3d40e656bd5ae72f1beef5dae8f51a1771e2/plugins/platforms/telegram/inline_picker.py#L45-L104'><code>collect_inline_catalog()</code> keeps core command and skill collection failures isolated rather than failing the whole handler.</a></li>
</ul>

<h3>Suggested contract</h3>

<p>Please make the authorization boundary explicit in a regression test: an allowlisted/pairing-authorized user querying inline mode from an unrelated chat should receive the same catalog as the supported private-chat identity path, while a user authorized only by a chat-specific rule must not gain access merely because <code>chat_id</code> is synthesized from <code>user_id</code>. Inline queries carry a user but no source chat, so this distinction is security-relevant.</p>

<p>Also cover the catalog failure branches: if core command collection fails or skill collection fails, the handler should still return a truthful partial/empty result and emit an observable diagnostic without exposing installed skill names. The existing tests cover authorized/unauthorized answers, pagination, and answer failure; an explicit cross-chat authorization case would close the remaining boundary.</p>

<p><strong>Scope:</strong> this review is based on Hermes PR #98317, its stated head revision, and the linked Hermes source/tests. The DeepSeek Harness handbook is an independent community project and is not Hermes documentation or a direct fix, so it is intentionally not linked here.</p>

PR #98319, opened 2026-08-30
feat(browser): add browser.real_profile_headed toggle to hide the real-profile window

### LINKED ISSUE #98311: feat: background/hidden mode for the real-profile browser
### Summary

When Hermes drives the **real-profile browser** (launches the installed browser — e.g. Brave — with `--user-data-dir` so the user's real login session is used), the browser window opens **visibly and steals focus**. There's no way to run that path in the background so the user can keep working in Hermes/other apps while the agent posts/navigates.

### Current state (verified in code)

- `hermes_cli/browser_connect.py` `launch_chrome_debug()` launches the installed browser with `--remote-debugging-port` + `--user-data-dir` (the real profile), and **no `--headless`** and **no hide-window flag** — so it's visible by design.
- The built-in headless browser (`tools/browser_tool.py`, `browser.headed: false` default) runs invisibly, but uses a **separate throwaway profile** — the user's real login (e.g. X/Twitter) is not present there.

So today it's a tradeoff: **real logged-in profile but visible window**, **or hidden window but no real login**. There's no option to get both.

### Requested

A way to run the real-profile browser in the **background** (hidden window / not stealing focus), while still using the real `--user-data-dir` profile. Likely shapes:

- A `--headless` option for the real-profile path (Chromium-family `--headless=new` still uses the user-data-dir), and/or
- A config toggle like `browser.real_profile_headed: false` (mirroring the existing `browser.headed`), and/or
- On Windows, a `SW_HIDE`/`STARTUPINFO.wShowWindow` flag on the real-profile launch (the built-in path already uses `windows_hide_flags()` — see `tools/browser_tool.py`).

### Use case

Agent-driven real-account automation (posting, checking a logged-in site) that should not interrupt the user's screen.

### Notes

This is a feature addition, not a bug — the current visible behavior is deliberate (real profile needs a real window today). Happy to help implement if maintainers want it.

chelsealong:
## Summary

Closes #98311.

`browser.use_real_profile` drives the user's real logged-in browser profile, but the launch always opens a **visible window that steals focus**, with no way to run it in the background. This adds a config toggle to hide it.

## Root cause

`tools/browser_tool.py::_real_profile_cdp()` launches the real Chrome/Edge/Brave/Chromium binary on a snapshot of the user's profile with `--no-startup-window`, then has agent-browser attach and run `open about:blank` over CDP. That `open` call is what actually creates the visible page/window — `--no-startup-window` only suppresses the window Chrome would otherwise open *before* that attach step. There was no equivalent of the built-in headless browser's `browser.headed` for this path, so the real-profile browser could never run hidden even though `--headless=new` was already used (and known to work, per an existing comment) for display-less Linux hosts.

## Fix

- `hermes_cli/config_defaults.py`: add `browser.real_profile_headed` (default `True`, preserving current visible-by-design behavior).
- `tools/browser_tool.py`: add `_real_profile_headed()` (mirrors the existing `_use_real_profile()` — reads the config on every call, no restart needed) and fold it into the existing headless-launch condition, so `browser.real_profile_headed: false` passes `--headless=new` to the real Chrome launch. This is the same flag already used for display-less Linux; the existing comment already establishes that Chrome's *new* headless mode (unlike legacy `--headless`) shares the profile's normal cookie store, so real-profile auth keeps loading.
- `website/docs/user-guide/features/browser.md`: document the new toggle next to the existing real-profile-browsing docs.

This is the "config toggle... mirroring the existing `browser.headed`" shape suggested in the issue. It does not add a Windows `SW_HIDE`/`STARTUPINFO` flag (the other shape mentioned) — `--headless=new` prevents a window from being created at all on every platform, which covers the reported problem without extra platform-specific code.

## Test plan

Added two tests to `tests/tools/test_browser_real_profile.py`:
- `test_hidden_mode_launches_chrome_headless` — with `browser.real_profile_headed` mocked to `False`, asserts `--headless=new` is present in the argv used to launch the real Chrome binary.
- `test_headed_default_keeps_chrome_visible` — with the toggle mocked to `True` (the default), asserts `--headless=new` is absent, so existing visible behavior is unchanged.

Verified both fail without the fix (reverted just the two source files with the tests still in place):

```
$ git stash push -- tools/browser_tool.py hermes_cli/config_defaults.py
$ scripts/run_tests.sh tests/tools/test_browser_real_profile.py -k "test_hidden_mode_launches_chrome_headless or test_headed_default_keeps_chrome_visible" -q
...
E   AttributeError: <module 'tools.browser_tool' ...> does not have the attribute '_real_profile_headed'
2 failed, 77 deselected
$ git stash pop
```

With the fix restored, the full file and related suites pass:

```
$ scripts/run_tests.sh tests/tools/test_browser_real_profile.py -q
=== Summary: 1 files, 79 tests passed, 0 failed (100% complete) in 5.1s (8 workers) ===

$ scripts/run_tests.sh tests/tools/test_browser_*.py tests/hermes_cli/test_config.py -q
=== Summary: 45 files, 722 tests passed, 0 failed, 4 skipped (100% complete) in 54.6s (8 workers) ===

$ scripts/run_tests.sh tests/hermes_cli/test_web_server.py -k schema -q
=== Summary: 1 files, 12 tests passed, 0 failed (100% complete) in 3.0s (8 workers) ===

$ uv run ruff check tools/browser_tool.py hermes_cli/config_defaults.py tests/tools/test_browser_real_profile.py
All checks passed!
```

## Platforms tested

Linux (GitHub-hosted runner), Python 3.11 (repo venv). No platform-specific code paths added — the fix is a single extra `--headless=new` on the existing Chrome launch, gated by the new config flag instead of only by "no display detected".

## AI assistance disclosure

This PR was written with AI assistance (Claude/Claude Code), with the diff reviewed and all listed tests run and verified before pushing.

### DIFF
--- a/hermes_cli/config_defaults.py
+++ b/hermes_cli/config_defaults.py
@@ -619,6 +619,13 @@ DEFAULT_CONFIG = {
         # naming a directory that doesn't exist FAILS CLOSED with a fixable
         # message rather than falling back to last-used.
         "real_profile_pin": "",
+        # Whether the real-profile browser opens a visible window. Default
+        # True preserves today's behavior (a real window, so a real login
+        # session is usable interactively). Set False to run it hidden —
+        # mirrors "headed" above, but scoped to the real-profile launch path
+        # since that one drives the user's actual logins rather than a
+        # throwaway profile.
+        "real_profile_headed": True,
         "allow_unsafe_evaluate": False,  # Legacy override: when true, browser_console(expression=...) bypasses the restrict_evaluate denylist entirely
         "restrict_evaluate": False,  # Opt-in denylist blocking sensitive JS primitives (cookies/storage/clipboard/network/form values) in browser_console(expression=...)
         # CDP supervisor — dialog + frame detection via a persistent WebSocket.
--- a/tests/tools/test_browser_real_profile.py
+++ b/tests/tools/test_browser_real_profile.py
@@ -361,6 +361,20 @@ class TestRealProfileCdpLaunch:
         assert "--headless=new" not in captured["argv"]
         self._reset()
 
+    def test_headed_string_false_is_hidden(self):
+        """A string "false" (e.g. from a generated/migrated config) must not
+        be treated as Python's truthy non-empty string."""
+        import tools.browser_tool as bt
+        with patch("hermes_cli.config.read_raw_config",
+                   return_value={"browser": {"real_profile_headed": "false"}}):
+            assert bt._real_profile_headed() is False
+
+    def test_headed_string_true_stays_headed(self):
+        import tools.browser_tool as bt
+        with patch("hermes_cli.config.read_raw_config",
+                   return_value={"browser": {"real_profile_headed": "true"}}):
+            assert bt._real_profile_headed() is True
+
     def test_reuses_only_session_on_our_copy_dir(self, tmp_path):
         """A live session on a DIFFERENT dir (stale/throwaway) is closed, not reused."""
         import tools.browser_tool as bt
--- a/tests/tools/test_browser_real_profile.py
+++ b/tests/tools/test_browser_real_profile.py
@@ -290,6 +290,77 @@ class TestRealProfileCdpLaunch:
         assert "--cdp" in captured["argv"]
         self._reset()
 
+    def test_hidden_mode_launches_chrome_headless(self, tmp_path):
+        """browser.real_profile_headed: false hides the real-profile window.
+
+        Unlike test_launch_never_passes_headless (which checks the
+        agent-browser ATTACH argv), this checks the argv Popen'd for the
+        real Chrome binary itself — that is where a visible window would
+        come from.
+        """
+        import tools.browser_tool as bt
+        self._reset()
+        proc = Mock(return_value=None, returncode=0, stdout="", stderr="")
+        captured = {}
+
+        class FakeChrome:
+            def poll(self):
+                return None
+
+        def fake_popen(argv, **kw):
+            captured["argv"] = argv
+            (tmp_path / "DevToolsActivePort").write_text("41000\n/devtools/browser/x\n")
+            return FakeChrome()
+
+        with patch.object(bt, "_use_real_profile", return_value=True), \
+             patch.object(bt, "_real_profile_headed", return_value=False), \
+             patch("hermes_cli.browser_connect.detect_default_chromium", return_value="chrome"), \
+             patch("hermes_cli.browser_connect.snapshot_real_profile", return_value=(str(tmp_path), None)), \
+             patch("hermes_cli.browser_connect.chromium_executable", return_value="/usr/bin/chrome"), \
+             patch.object(bt.subprocess, "Popen", side_effect=fake_popen), \
+             patch.object(bt, "_agent_browser_get_cdp",
+                          side_effect=[None, "http://127.0.0.1:41000"]), \
+             patch.object(bt, "_find_agent_browser", return_value="/usr/bin/agent-browser"), \
+             patch.object(bt.subprocess, "run", return_value=proc), \
+             patch("sys.platform", "darwin"):
+            cdp, err = bt._real_profile_cdp()
+        assert err is None
+        assert cdp == "http://127.0.0.1:41000"
+        assert "--headless=new" in captured["argv"]
+        self._reset()
+
+    def test_headed_default_keeps_chrome_visible(self, tmp_path):
+        """Without the opt-out, the real-profile Chrome launch stays visible."""
+        import tools.browser_tool as bt
+        self._reset()
+        proc = Mock(return_value=None, returncode=0, stdout="", stderr="")
+        captured = {}
+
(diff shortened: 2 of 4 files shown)


liyangbing:
<h3>Review</h3>

<p>The implementation is focused and the two launch tests cover the primary regression: <code>--headless=new</code> is added only for the hidden real-profile path, while the default remains headed.</p>

<h3>Configuration boundary</h3>

<p><a href='https://github.com/NousResearch/hermes-agent/blob/af14cfef4aff688d008a79122878d91419583ab4/tools/browser_tool.py#L1451-L1470'><code>_real_profile_headed()</code> returns <code>bool(browser_cfg.get(...))</code></a>. If <code>read_raw_config()</code> can yield a string value (for example a generated or migrated config containing <code>real_profile_headed: &quot;false&quot;</code>), Python treats that non-empty string as <code>True</code>, silently leaving the window visible. Please either normalize accepted string booleans in this helper or reject them with a clear configuration error, and add a regression for the chosen contract. The unquoted YAML boolean in the new example should remain covered as the normal path.</p>

<h3>Documentation wording</h3>

<p>The new browser guide says “When you turn the toggle back off, Hermes deletes the snapshot store,” but the preceding toggle is <code>real_profile_headed</code>; snapshot deletion is controlled by <code>use_real_profile</code>. Please make that sentence name the correct setting so users do not believe switching from hidden to headed deletes their copied profile.</p>

<p><strong>Scope:</strong> this review is based on Hermes PR #98319, its stated head revision, and the linked Hermes source/tests. The DeepSeek Harness handbook is an independent community project and is not Hermes documentation or a direct fix, so it is intentionally not linked here.</p>

chelsealong:
Addressed both points in ad3366b10:

- \`_real_profile_headed()\` now uses the shared \`is_truthy_value()\` helper (already used elsewhere in this codebase) instead of \`bool()\`, so a string \`"false"\` no longer resolves to Python's truthy non-empty string. Added \`test_headed_string_false_is_hidden\` / \`test_headed_string_true_stays_headed\`, verified both fail against the prior \`bool()\` implementation.
- Fixed the doc wording: snapshot-store deletion is tied to \`use_real_profile\`, not \`real_profile_headed\` — the guide now names the correct setting.

\`scripts/run_tests.sh tests/tools/test_browser_*.py tests/hermes_cli/test_config.py -q\` → 724 passed, 0 failed. \`ruff check\` clean on all touched files.

DinGood:
Confirming this toggle is not just an UX nicety — on Windows it is the difference between working and permanently hung local browsing.**Env:** Hermes Agent v0.21.0 (git install, desktop app), Windows 10/11, Google Chrome 152 (current stable), `browser.use_real_profile: true`, harness 0.1.13.**Symptom:** Every `browser_exec local=true` call hangs until the full tool timeout (420s), even after the user completes BOTH consent stages (chrome://inspect checkbox + Allow popup) on their foreground Chrome.**Root cause observed live:**- `_resolve_real_profile_cdp` points the harness (BU_CDP_* env) at agent-browser's real-profile snapshot browser, launched headless (per #98323's behavior).- Process enumeration: 11 chrome.exe processes with `agent-browser-chrome-*` temp user-data-dir, zero MainWindowTitle — no window exists to host the consent dialog.- Chrome ≥136 requires a visible-window user click to allow CDP on a real profile (#102986's own body calls this "blocked outright on Chrome >= 136").- Result: the snapshot's CDP port LISTENs but `GET /json/version` returns an empty reply forever — nothing can ever consent a headless window. Wedged daemons also survive across retries (#100998 / #100855 territory).**Workaround in use:** piping Python to a standalone `browser-harness` against the consented foreground Chrome works at 2s/page — proving the browser and Chrome-side consent are fine; only the headless snapshot route deadlocks.This PR (a headed option for real-profile sessions) is the minimal unblock for Windows real-profile browsing. Please consider re-prioritizing from P3 — on Windows this lane currently has a 0% success rate without it.---*AI-assistance disclosure: this comment was drafted with the help of Hermes Agent (the AI assistant the reporter runs on this machine), based on live diagnostics it performed on the Windows host — including the source-code reading, process enumeration, and CDP probe evidence cited above. The reporter reviewed the content and is submitting it under their own account.*

alt-glitch:

Duplicate of merged #98323 — real-profile browsing now launches `--headless=new` by default on every platform, and the visible window is opt-in via the existing `browser.headed` / `AGENT_BROWSER_HEADED` toggle (now honored for the real-profile path too). That supersedes the separate `browser.real_profile_headed` knob added here; #98311 is closed. @DinGood's Windows report (hang until 420s timeout while the foreground Chrome waits on consent) should be covered by the merged default — if it still reproduces on current main, please open a fresh issue with that log.

chelsealong:
Closing as superseded: the linked issue #98311 was closed as completed on 2026-08-30, so the underlying bug is fixed on main. I checked the current tree and main's fix does not use this branch's code — it was resolved a different way — so there is nothing left for this PR to contribute. Closing rather than leaving a stale branch open against a resolved issue.

PR #98320, opened 2026-08-30
feat(usage): add Kimi for Coding account quota to fetch_account_usage

DanielLSM:
## What

- Adds a Kimi for Coding account-quota probe to `fetch_account_usage`: it GETs `<base>/usages` with the resolved `kimi-coding` credential and maps the response into `AccountUsageWindow`s (the weekly window plus each entry in `limits`, with human labels derived from the window duration, e.g. `Current 5h`, `Current week`).
- The `/usages` API omits `used` for buckets with zero consumption; consumption is then derived from `limit - remaining` so untouched windows still render instead of silently vanishing.
- `_format_reset` now formats reset times in the configured Hermes timezone (falls back to system local when unavailable).

## Why

Kimi for Coding subscriptions enforce weekly and shorter rolling quota windows, but `fetch_account_usage` returned nothing for the provider, so there was no visibility into remaining budget from Hermes.

## Tests

- `python -m pytest tests/test_account_usage.py -q` → 8 passed, including new coverage for window labeling, zero-`used` derivation, and snapshot mapping.

No config changes, no new dependencies.

### DIFF
--- a/agent/account_usage.py
+++ b/agent/account_usage.py
@@ -75,7 +75,13 @@ def _parse_dt(value: Any) -> Optional[datetime]:
 def _format_reset(dt: Optional[datetime]) -> str:
     if not dt:
         return "unknown"
-    local_dt = dt.astimezone()
+    try:
+        from hermes_time import get_timezone
+
+        tz = get_timezone()
+    except Exception:
+        tz = None
+    local_dt = dt.astimezone(tz) if tz is not None else dt.astimezone()
     delta = dt - _utc_now()
     total_seconds = int(delta.total_seconds())
     if total_seconds <= 0:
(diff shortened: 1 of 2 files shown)


liyangbing:
<h3>Review</h3>

<p>The Kimi usage mapping is localized to <a href='https://github.com/NousResearch/hermes-agent/blob/c5dc6f722fa4f1cec4a9432c671c51ce6ce12585/agent/account_usage.py#L887-L1031'><code>agent/account_usage.py</code></a>, and the current tests establish the normal weekly/window mapping, aliases, missing credentials, and malformed numeric input.</p>

<h3>Missing-zero-consumption regression</h3>

<p>The implementation explicitly supports the API shape where <code>used</code> is omitted and derives it from <code>limit - remaining</code>, but the test payload always supplies <code>used</code>. Please add a fixture with an untouched bucket such as <code>{&quot;limit&quot;: &quot;100&quot;, &quot;remaining&quot;: &quot;100&quot;}</code> and assert a rendered <code>0% used / 100% remaining</code> window. Also cover a bucket where <code>used</code> is absent and <code>remaining</code> is malformed: it should be skipped without hiding valid sibling windows.</p>

<h3>Alias contract</h3>

<p><a href='https://github.com/NousResearch/hermes-agent/blob/c5dc6f722fa4f1cec4a9432c671c51ce6ce12585/agent/account_usage.py#L887-L887'><code>KIMI_CODING_PROVIDERS</code></a> includes <code>kimi-coding</code>, <code>kimi-code</code>, <code>kimi</code>, <code>moonshot</code>, and <code>moonshot-cn</code>, while the alias test covers only the first four variants other than the canonical name. Please parameterize over the complete set so a future alias change cannot silently diverge from dispatch behavior.</p>

<p>For the API contract, retain the existing behavior that invalid or non-positive limits do not create a usage window, and assert that a malformed top-level <code>usage</code> object does not prevent valid entries in <code>limits</code> from rendering. These are the failure-isolation branches in <code>_kimi_usage_percent()</code> and <code>_fetch_kimi_account_usage()</code>, not provider-specific assumptions.</p>

<p><strong>Scope:</strong> this review is based on Hermes PR #98320, its stated head revision, and the linked Hermes source/tests. The DeepSeek Harness handbook is an independent community project and is not Hermes documentation or a direct fix, so it is intentionally not linked here.</p>

PR #98323, opened 2026-08-30
fix(browser): real-profile browsing runs headless — no focus-stealing window

teknium1:
## Summary
Real-profile browsing now runs **headless** — it drives the copied profile in the background with no visible window, instead of popping the user's real browser and stealing focus on every turn.

Root cause: the native launch (which bypasses agent-browser's own launcher to avoid `--use-mock-keychain` dropping keychain-encrypted cookies) only appended `--headless=new` when Linux had no `DISPLAY`/`WAYLAND_DISPLAY`. On any normal desktop that guard was false, so Chrome opened a visible, focus-grabbing window — the opposite of the feature's intent (drive a copy headlessly so the user can keep working).

## Changes
- `tools/browser_tool.py`: `_real_profile_cdp()` launches `--headless=new` by default on every platform. Opt in to a visible window via the existing `browser.headed` / `AGENT_BROWSER_HEADED` toggle (now honored for real-profile too, same as the rest of the browser stack); display-less hosts stay headless regardless so the launch doesn't die at startup.
- `tests/tools/test_browser_real_profile.py`: replaced the stale `test_launch_never_passes_headless` (built on a legacy-headless "separate cookie store" premise) with `test_launch_is_headless_and_agent_browser_attaches`, which positively asserts the chrome launch carries `--headless=new` by default while agent-browser still attaches via `--cdp`.
- `website/docs/user-guide/features/browser.md`: documents that real-profile browsing runs headless in the background, why logins still load (Chrome's new headless shares the cookie store), and the headed opt-in.

## Why headless is safe here
The cookie drop we guard against comes from `--use-mock-keychain` (which agent-browser's own launcher force-adds and we deliberately avoid), **not** from headless mode. Chrome's *new* headless (`--headless=new`) reads the profile's normal cookie store — unlike legacy `--headless` — so the copied auth/login state loads exactly as before.

## Validation
Live A/B on a real X seat (DISPLAY=:0), launching the real Chrome binary with the old vs new argv and counting mapped windows:

| | argv | mapped chrome windows | CDP port |
|---|---|---|---|
| Before (old guard) | no `--headless` | **1** (focus steal) | live |
| After (fix) | `--headless=new` | **0** | live |

Targeted tests: `tests/tools/test_browser_real_profile.py` + `test_browser_real_profile_pin.py` → 81 passed.

## Infographic

Real-Profile Browsing Runs Headless

### DIFF
--- a/tests/tools/test_browser_real_profile.py
+++ b/tests/tools/test_browser_real_profile.py
@@ -248,14 +248,22 @@ class TestRealProfileCdpLaunch:
         assert cdp == "http://127.0.0.1:41000"
         self._reset()
 
-    def test_launch_never_passes_headless(self, tmp_path):
-        """--headless would use a separate cookie store → 0 real cookies.
+    def test_launch_is_headless_and_agent_browser_attaches(self, tmp_path):
+        """Real-profile browsing runs headless (no focus-stealing window).
 
-        We launch the REAL Chrome binary
-        ourselves (no mock-keychain switches — those break macOS cookie
-        decryption) and agent-browser attaches via --cdp. So the agent-browser
-        argv must contain --cdp (attach, not launch) and must NOT contain
-        --headless or --profile (launch-mode switches).
+        Two argv paths are checked:
+
+        1. The REAL Chrome binary we launch ourselves (via Popen) MUST pass
+           ``--headless=new``. Real-profile browsing is a background
+           capability — a visible window that grabs focus every turn defeats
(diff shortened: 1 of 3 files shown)


alt-glitch:

Related: #98319 targets the same real-profile visibility request with a separate dedicated toggle while preserving visible-by-default behavior. This PR instead makes the existing headed toggle apply to real-profile browsing and changes the default to headless; maintainers should choose the intended default/configuration shape.

PR #98306, opened 2026-08-30
fix(todo): unversioned tool.start still merges after resume

Adolanium:
## What does this PR do?

#97815 taught the desktop to reject todo updates that have no revision. After that work landed (via salvage #98248), a `merge: true` `tool.start` after resume never patches the list.

Resume attaches those snapshots so the Tasks panel survives reconnect. An unused store serializes as `{todos: [], revision: 0}`, and the desktop treats a missing revision as stale. `tool.start` for a `merge: true` write has no revision, so the patch from #97811 never applies after resume.

Skip unused empty snapshots. Apply unversioned updates without moving the watermark so a later `todo.updated` can still win.

## Related Issue

Fixes #98304

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `tui_gateway/server.py`: `_normalize_todo_state` returns None for `{todos: [], revision: 0}`. An empty list at revision >= 1 is still a real clear.
- `apps/desktop/src/store/todos.ts`: unversioned `setSessionTodos` applies and does not change the watermark. Restore ignores unused empty snapshots.
- `apps/desktop/src/store/todos.test.ts` and `tests/tui_gateway/test_todo_state_events.py`: start merge after a revisioned snapshot, unused snapshot is not attached.

## How to Test

1. From `apps/desktop`: `npx vitest run src/store/todos.test.ts`
2. `python scripts/run_tests_parallel.py tests/tui_gateway/test_todo_state_events.py`
3. After `setSessionTodos(..., 5)`, an unversioned `setSessionTodos` must still update the list and leave revision at 5. `{todos: [], revision: 0}` must not stamp a watermark.

Ran on Windows 11: 13 vitest passed, 5 python tests passed. ruff clean on the python files.

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
[...]

PR #98326, opened 2026-08-30
fix(todo): unversioned tool.start merges again after desktop resume (#98304, salvage #98306)

### LINKED ISSUE #98304: [Bug]: Tasks panel skips merge:true todo start after resume
### Bug Description

#97815 taught the desktop to reject todo updates that have no revision. After that work landed (via salvage #98248), a `merge: true` `tool.start` after resume never patches the list.

Resume/activate can attach an unused todo store as `{todos: [], revision: 0}`. Live `todo` `tool.start` has no revision, so the update is dropped. A write that used to patch one row in place now waits until `tool.complete` / `todo.updated`. The 1/1 collapse from #97810 stays gone, but the start merge from #97811 never runs after a reconnect.

### Steps to Reproduce

1. Open a desktop session, resume or activate it (even one that never used `todo`).
2. Start a turn that already has a few todo items, then call `todo` with `merge: true` and one item marked completed or in_progress.
3. Watch the Tasks panel on `tool.start`, before `tool.complete`.

### Expected Behavior

The existing list stays. The patched row updates on `tool.start`. Complete still replaces with the full store.

### Actual Behavior

`setSessionTodos` gets a null revision from `tool.start` and returns early because a watermark is already set (0 from the empty resume snapshot, or a later live revision). The panel does not merge until the complete snapshot lands.

### Affected Component

- Other

N/A (desktop composer)

### Debug Report

N/A. Found by reading #97815 (salvaged in #98248) on top of #97811. Reproduced with unit tests: an unversioned `setSessionTodos` after a revisioned snapshot must still apply, and `{todos: [], revision: 0}` must not stamp a watermark.

### Operating System

Windows 11

### Python Version

3.11

### Hermes Version

main (includes #98248, salvage of #97815)

#97815 is the source. It added the resume snapshot attach and desktop revision arbitration. #98248 only salvaged and merged that core.

`tui_gateway/server.py` `_normalize_todo_state` treats an unused `TodoStore.snapshot()` (`todos: []`, `revision: 0`) as a real snapshot and `_attach_todo_state` puts it on resume/activate.

`apps/desktop/src/store/todos.ts` `acceptRevision` returns false when `revision == null` and a watermark already exists. `tool.start` never carries a revision. #97811 called `setSessionTodos(sessionId, todos)` with no revision on purpose.

### Proposed Fix (optional)

Do not attach unused empty snapshots. Apply unversioned updates without moving the watermark so a later `todo.updated` can still win.

teknium1:
## Summary

`merge: true` todo `tool.start` patches apply again after a desktop resume/reconnect — the regression #98248 introduced against the #97811 behavior it salvaged.

Root cause: resume/activate attaches the agent's TodoStore snapshot, and an unused store serializes as `{todos: [], revision: 0}`. The desktop treated that as a real revision and stamped a per-session watermark; unversioned updates (which is what `tool.start` merge patches are) were then rejected forever, so the panel froze until `tool.complete`.

Salvage of #98306 by @Adolanium (commit preserved with authorship). Fixes #98304.

## Changes

- `tui_gateway/server.py`: `_normalize_todo_state` returns None for the unused `{todos: [], revision: 0}` placeholder, so resume never attaches it; an empty list at revision >= 1 is still a real clear
- `apps/desktop/src/store/todos.ts`: unversioned `setSessionTodos` applies without moving the watermark (a later `todo.updated` still wins); restore ignores unused empty snapshots
- Regression tests in `todos.test.ts` + `test_todo_state_events.py`

## Validation

| Check | Result |
|---|---|
| Sabotage run (fix reverted to main's code) | 2 vitest + 1 pytest new tests FAIL as expected |
| `apps/desktop` vitest `todos.test.ts` + `todo-cleanup.test.tsx` | 17/17 pass |
| `tests/tui_gateway/test_todo_state_events.py` | 5/5 pass |
| `tsc --noEmit` | clean |
| `scripts/audit_pr_attribution.py --fix` | all emails mapped |

## Credits

- @Adolanium — #98306: full fix + regression tests (cherry-picked, authorship preserved)

## Infographic

Unversioned tool.start merges again after resume

### DIFF
--- a/apps/desktop/src/store/todos.test.ts
+++ b/apps/desktop/src/store/todos.test.ts
@@ -134,4 +134,22 @@ describe('revisioned snapshots', () => {
     restoreSessionTodosFromSnapshot('s1', snapshot, true)
     expect($todosBySession.get().s1?.[0]?.id).toBe('active')
   })
+
+  it('applies an unversioned update after a revisioned snapshot (tool.start merge)', () => {
+    setSessionTodos('s1', [todo('a', 'pending'), todo('b', 'pending')], 5)
+    setSessionTodos('s1', [todo('a', 'completed'), todo('b', 'pending')])
+
+    expect($todosBySession.get().s1?.[0]?.status).toBe('completed')
+    expect($todoRevisionsBySession.get().s1).toBe(5)
+  })
+
+  it('does not stamp a watermark from an unused empty snapshot', () => {
+    restoreSessionTodosFromSnapshot('s1', { revision: 0, todos: [] }, true)
+
+    expect($todosBySession.get().s1).toBeUndefined()
+    expect($todoRevisionsBySession.get().s1).toBeUndefined()
+
+    setSessionTodos('s1', [todo('a', 'in_progress')])
+    expect($todosBySession.get().s1?.[0]?.id).toBe('a')
+  })
 })
--- a/apps/desktop/src/store/todos.ts
+++ b/apps/desktop/src/store/todos.ts
@@ -74,8 +74,10 @@ function acceptRevision(sid: string, revision?: null | number): boolean {
   const revisions = $todoRevisionsBySession.get()
   const current = revisions[sid]
 
+  // tool.start has no revision. Apply the merge locally and leave the
+  // watermark alone so a later todo.updated / tool.complete can still win.
   if (revision == null) {
-    return current == null
+    return true
   }
 
   if (current != null && revision < current) {
@@ -156,6 +158,14 @@ export function restoreSessionTodosFromSnapshot(sid: string, snapshot: unknown, 
   }
 
   const revision = parseTodoRevision(snapshot)
+
+  // An unused store serializes as {todos: [], revision: 0}. That is not a
+  // real snapshot. Applying it would stamp watermark 0 and leave an empty
+  // list in the map.
+  if (todos.length === 0 && (revision == null || revision === 0)) {
+    return
+  }
+
   const visible = running ? todos : todosForHydration(todos)
 
   if (visible !== null) {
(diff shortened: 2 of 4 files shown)


liyangbing:
<h3>Review</h3>

<p>The revision-0 placeholder fix is well targeted: <a href='https://github.com/NousResearch/hermes-agent/blob/fd719673f21ce60123c4c49039d95b2706cabc29/tui_gateway/server.py#L7645-L7660'><code>_normalize_todo_state()</code> now omits the unused store</a>, while a nonzero empty snapshot remains a real clear. The desktop tests also pin that distinction.</p>

<h3>Out-of-order update contract</h3>

<p><a href='https://github.com/NousResearch/hermes-agent/blob/fd719673f21ce60123c4c49039d95b2706cabc29/apps/desktop/src/store/todos.ts#L74-L86'><code>acceptRevision()</code> now returns <code>true</code> for every unversioned update</a>. That is correct for a <code>tool.start</code> merge arriving after a revisioned snapshot, but it also permits a delayed unversioned update to be applied after a newer revisioned <code>todo.updated</code> or <code>tool.complete</code>. The comment says the later revisioned event can win, but network/event ordering can be the reverse.</p>

<p>Please define and test the ordering contract explicitly: either unversioned <code>tool.start</code> is allowed to update only fields that cannot regress a revisioned terminal state, or the client needs an event-generation/sequence fence in addition to the snapshot watermark. A late <code>tool.start</code> must not resurrect completed items, remove a newer item, or move a terminal session back to an older visible state.</p>

<h3>Regression matrix</h3>

<p>Cover revisioned snapshot → unversioned start, unversioned start → newer revisioned update, and newer revisioned update → delayed unversioned start. Include an empty revision-0 snapshot, a nonzero revision empty clear, a completed todo, and a newly added todo. Assert both the visible list and the retained watermark after each ordering.</p>

<p><strong>Scope:</strong> this review is based on Hermes PR #98326, its stated head revision, and the linked Hermes source/tests. The DeepSeek Harness handbook is an independent community project and is not Hermes documentation or a direct fix, so it is intentionally not linked here.</p>

PR #94005, opened 2026-08-24
fix(display): clamp spinner frames to terminal width

Ahmett101:
Summary
`KawaiiSpinner` wrote every TTY animation frame with `
` while allowing the rendered message to exceed the terminal width. In narrow PowerShell/Windows consoles a long tool preview can wrap, and carriage return then rewrites the wrapped continuation row instead of the original spinner row, flooding scrollback. This clamps spinner frames to the captured terminal width minus one autowrap cell, tracks padding by terminal cell width, and keeps wide-character frames within the same boundary.

Changes
agent/display.py: adds terminal-cell width helpers, resolves the captured output stream width safely, clamps spinner frames before the autowrap column, and pads using display cells rather than raw string length.
tests/agent/test_display.py: adds regression coverage for long PowerShell-style paths and wide-character spinner content.

How to Test
/data/data/com.termux/files/home/.hermes/hermes-agent/venv/bin/python -m pytest tests/agent/test_display.py -q
# ✅ 25/25 passed
ruff check agent/display.py tests/agent/test_display.py
# ✅ passed
git diff --check
# ✅ passed
python scripts/check-windows-footguns.py --diff upstream/main
# ✅ passed
scripts/run_tests.sh
# ⚠️ blocked by Termux runner PermissionError(13, 'Permission denied') environment issue; targeted tests above passed

Checklist
- [x] Tests pass — 25/25 targeted
- [x] Follows Conventional Commits
- [x] Changes scoped to this fix only
- [x] Cross-platform impact assessed (Linux / macOS / WSL2 / Windows / Termux)
- [x] profile-safe paths used — no hardcoded ~/.hermes
- [x] .env not used for non-credential settings (behavioral settings → config.yaml)

Risk & Impact
Low. The change only shortens transient spinner frames before terminal autowrap; non-TTY output remains unchanged and wide-character padding is more accurate.
Type: Bug fix
Closes: #93999

Issue #93999, opened 2026-08-24
Bug: KawaiiSpinner floods terminal when message exceeds terminal width (PowerShell, narrow consoles)


DarkLeopard:
# Bug: KawaiiSpinner floods terminal when message exceeds terminal width (PowerShell, narrow consoles)

## Summary

`KawaiiSpinner._animate()` writes each animation frame with `\r` to overwrite the previous line in place. When the spinner message (tool name + preview) is longer than the terminal width, the line wraps. `\r` (carriage return) only returns to the start of the *current visual line* (the wrapped second line), not the original first line. Each animation frame (0.12s interval) lands on a new line, flooding the terminal with spinner frames.

On a standard 80-column PowerShell window, a tool preview like `Running head -5 "C:/Users/Leo/.hermes/profiles/dfs-lead/skills/software..."` easily exceeds 80 chars, causing the screen to fill with dozens of spinner lines for a single tool call.

## Root Cause

`agent/display.py`, `KawaiiSpinner._animate()`:

```python
line = f"  {frame} {self.message} ({elapsed:.1f}s)"
pad = max(self.last_line_len - len(line), 0)
self._write(f"\r{line}{' ' * pad}", end='', flush=True)
```

`self.message` includes the full tool preview (via `build_tool_label` → `build_tool_preview`). The global `_tool_preview_max_len` defaults to `0` (unlimited), so long file paths are passed through untruncated. When `len(line) > terminal_width`, the line wraps and `\r` cannot return to the original line start.

## Reproduction

1. Run Hermes CLI in PowerShell (or any 80-column terminal)
2. Trigger a tool call with a long preview, e.g. `terminal` with a long command path, or `read_file` on a deeply nested path
3. Observe the spinner: each animation frame (⠋⠙⠹⠸⠼⠴⠦) appears on a new line instead of overwriting in place

Before (e.g. when stdout was a pipe, `_is_tty=False`): one line `[tool] {message}`, no animation — clean.
After (PowerShell TTY, `_is_tty=True`): animation runs, `\r` fails on wrapped lines — screen floods.

## Fix

Truncate the spinner line to terminal width before writing:

```python
try:
    import shutil as _sh
    _cols = _sh.get_terminal_size(self._out.fileno() if hasattr(self._out, 'fileno') else 0).columns
except (ValueError, OSError):
    _cols = 80
if _cols > 0 and len(line) > _cols:
    line = line[:_cols - 1]
```

### Files changed

[...]

PR #98327, opened 2026-08-30
fix(display): skip \r spinner animation when print_fn is set

SayHell0W0rld:
Fixes the two residual paths in #93999 that #94005 does not cover.

## Problem

`KawaiiSpinner._animate()` runs a `\r`-based animation in all cases. When a `print_fn` is configured:

1. **Flood** — `_write()` routes through `print_fn` → `_cprint` (cli_agent_setup_mixin.py:558) → prompt_toolkit `print_formatted_text`, which appends a newline per call. Every spinner frame lands on its own terminal line, flooding the scrollback — independent of frame width, so #94005's width clamp does not help.
2. **Flicker** — the CLI TUI already drives spinner display via `_render_spinner_text()` → `get_spinner_text` widget + `app.invalidate()` (cli.py:19812). A concurrent `\r` animation fights the widget for the status line → visual flicker on Windows conhost.

`stop()` is also guarded: when `print_fn` is set there is no animation line to clear, and routing `\r{blanks}\r` through `print_fn` would emit a stray line of spaces.

## Fix

`_animate()` returns early (sleep loop) when `self._print_fn is not None`, same rationale as the existing StdoutProxy branch. `stop()` skips the `\r` clear when `print_fn` is set. The completion message still flows through `print_fn`.

## Tests

- `test_kawaii_spinner_skips_animation_when_print_fn_set`: print_fn receives no `\r`-prefixed frames; completion message still delivered.
- `test_kawaii_spinner_animate_runs_when_no_print_fn`: without print_fn, `\r` animation still runs on a tty.

`pytest tests/agent/test_display.py` → 25 passed. ruff clean.

Complementary to #94005 (references the same issue, different code path).

### DIFF
--- a/agent/display.py
+++ b/agent/display.py
@@ -1220,6 +1220,21 @@ class KawaiiSpinner:
                 time.sleep(0.5)
             return
 
+        # When a print_fn is configured, the \r-based animation must not run:
+        #   * CLI routes output through _cprint (cli_agent_setup_mixin.py)
+        #     → prompt_toolkit print_formatted_text, which appends a newline
+        #     per call — so every frame would land on its own line and flood
+        #     the terminal, regardless of frame width.
+        #   * The CLI TUI already renders spinner state via a dedicated
+        #     widget (_spinner_text → _render_spinner_text → get_spinner_text),
+        #     so a \r animation here would also fight the widget for the same
+        #     status line (visual flicker on Windows conhost).
+        # Same rationale as the StdoutProxy branch below.
+        if self._print_fn is not None:
+            while self.running:
+                time.sleep(0.1)
+            return
+
         # When running inside prompt_toolkit's patch_stdout context the CLI
         # renders spinner state via a dedicated TUI widget (_spinner_text).
         # Driving a \r-based animation here too causes visual overdraw: the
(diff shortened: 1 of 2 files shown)


liyangbing:
<h3>Review</h3>

<p>The change correctly separates the <code>print_fn</code> path from carriage-return animation: <a href='https://github.com/NousResearch/hermes-agent/blob/8a9bd0016fbb7e1f4cd362611a8f831b09f871cd/agent/display.py#L1220-L1240'><code>_animate()</code> waits without emitting frames</a>, and <a href='https://github.com/NousResearch/hermes-agent/blob/8a9bd0016fbb7e1f4cd362611a8f831b09f871cd/agent/display.py#L1301-L1310'><code>stop()</code> skips the tty clear</a>.</p>

<h3>Lifecycle and output contract</h3>

<p>The new test proves that no <code>\r</code>-prefixed frame reaches <code>print_fn</code>, but it does not prove that the wait loop has exited after <code>stop()</code>. Please assert that the animation thread is no longer alive after stopping, and cover an immediate <code>start()</code> → <code>stop()</code> race as well as a stop after several wait iterations. This protects the new branch from leaking one daemon thread per task.</p>

<p>Also assert that <code>stop(final_message=...)</code> sends only the completion message through <code>print_fn</code>: no blank-line clearing payload, no spaces-only call, and no duplicate completion when <code>stop()</code> is called twice. The production guard is correct, but these assertions pin the reason for skipping the tty clear rather than only checking the absence of a carriage return.</p>

<h3>Regression matrix</h3>

<p>Cover <code>print_fn=None</code> on a tty, <code>print_fn</code> on a tty, non-tty output, immediate stop, repeated stop, and completion with an empty or absent final message. Keep the existing positive assertion that the ordinary tty path still emits carriage-return frames.</p>

<p><strong>Scope:</strong> this review is based on Hermes PR #98327, its stated head revision, and the linked Hermes source/tests. The DeepSeek Harness handbook is an independent community project and is not Hermes documentation or a direct fix, so it is intentionally not linked here.</p>

PR #43781, opened 2026-06-10
feat(discord): support threaded free-response channels

snapss-dev:
## Summary
- add `discord.free_response_threaded_channels` / `DISCORD_FREE_RESPONSE_THREADED_CHANNELS`
- let selected Discord channels bypass @mention gating while still using auto-created threads for top-level messages
- keep parent-channel matching for replies in threads under those channels

## Verification
- `/Users/snaps/.hermes/hermes-agent/venv/bin/python -m pytest tests/gateway/test_discord_channel_controls.py -q`
- `/Users/snaps/.hermes/hermes-agent/venv/bin/python -m py_compile plugins/platforms/discord/adapter.py`

PR #98328, opened 2026-08-30
fix(discord): restore auto-threading for threaded free-response lanes

marcelomediagroup:
## Summary
- add profile-scoped `discord.threaded_free_response_channels` for mention-free intake lanes that still create one task thread per parent message
- preserve thread-bound final/status routing and isolate parent voice mode
- carry the scopes through relay admission, env/YAML/scalar normalization, and missed-message recovery defaults
- document the canonical config and compatibility environment surface

## Behavior contracts
- ordinary free-response channels remain inline
- `no_thread_channels` retains precedence
- multiplexed profiles do not share `auto_thread`
- relay scope-only policy keeps `requireAddress: true` outside the listed scopes
- empty recovery-channel lists fall back to allowed/free/threaded scopes
- default compaction-notice suppression is unchanged

## Verification
- rebased onto current `origin/main` (`58523f284c`); range-diff reports the reviewed patch is identical
- focused final suite: 125 passed via `scripts/run_tests.sh`
- combined Discord/relay/config suites before rebase: 836 passed, 4 Windows-only skipped
- Ruff, Windows-footgun scan, and `git diff --check`: clean
- independent Standards and Spec reviews: no hard findings

No live gateway restart, Discord message, or voice/session mutation is included in this PR.

### DIFF
--- a/cli-config.yaml.example
+++ b/cli-config.yaml.example
@@ -1251,6 +1251,7 @@ platform_toolsets:
 #   require_mention: true            # Require @mention in server channels (default: true)
 #   auto_thread: true                # Auto-create thread on @mention (default: true)
 #   free_response_channels: ""       # Channel IDs where no mention is needed
+#   threaded_free_response_channels: []  # Mention-free lanes that still auto-thread
 #   reactions: true                  # Show processing reactions (default: true)
 #   history_backfill: true           # Recover missed channel messages on mention (default: true)
 #   history_backfill_limit: 50       # Max messages to scan backwards (default: 50)
--- a/docs/relay-connector-contract.md
+++ b/docs/relay-connector-contract.md
@@ -693,6 +693,7 @@ driven by the managed Portal / `hermes` CLI.
 
 The relevance layer (§7.1) is the per-tenant parity for the gateway's own
 behaviour knobs (`require_mention`, `free_response_channels`,
+`threaded_free_response_channels`,
 `{PLATFORM}_ALLOW_BOTS`). So the **same** behaviour governs relay delivery, the
 gateway projects those knobs into a **platform-agnostic** policy and POSTs it to
 `POST /relay/policy` at boot (after its per-gateway secret is resolved).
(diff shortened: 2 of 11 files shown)


alt-glitch:

Related: #43781 implements the same Discord threaded-free-response setting at the adapter layer. This PR additionally carries the setting through relay policy and recovery/config compatibility; maintainers should choose whether that broader scope is preferred.

liyangbing:
<h3>Review</h3>

<p>The implementation correctly separates mention admission from thread creation: <a href='https://github.com/NousResearch/hermes-agent/blob/3ce71a9418fa2256dc6f0143c035228f411fcc9a/plugins/platforms/discord/adapter.py#L8184-L8238'><code>threaded_free_response_channels</code> joins the mention-free set, while <code>no_thread_channels</code> remains a skip condition</a>. The relay and recovery paths also include the new scope.</p>

<h3>Precedence contract</h3>

<p>The behavior needs one explicit precedence rule. The same branch still gates creation on <a href='https://github.com/NousResearch/hermes-agent/blob/3ce71a9418fa2256dc6f0143c035228f411fcc9a/plugins/platforms/discord/adapter.py#L6743-L6775'><code>_discord_auto_thread_enabled()</code></a>. Therefore a channel listed in <code>threaded_free_response_channels</code> does not create a thread when the profile's global <code>auto_thread</code> is <code>false</code>, despite the documentation saying that the new setting “still auto-creates” a thread.</p>

<p>Please either document that <code>auto_thread: false</code> is the higher-precedence kill switch, or make the per-channel setting intentionally override it. Whichever contract is intended, test the combination explicitly, including a multiplexed second profile whose <code>auto_thread</code> value differs. A silent inline fallback is particularly confusing because mention admission still succeeds.</p>

<h3>Regression matrix</h3>

<p>Cover threaded-free with <code>auto_thread: true</code>, threaded-free with <code>auto_thread: false</code>, overlap with <code>no_thread_channels</code>, overlap with ordinary <code>free_response_channels</code>, parent-channel versus existing-thread routing, and recovery of an unmentioned parent message. Assert both whether <code>_auto_create_thread()</code> was called and the resulting event/send scope.</p>

<p><strong>Scope:</strong> this review is based on Hermes PR #98328, its stated head revision, and the linked Hermes source/tests. The DeepSeek Harness handbook is an independent community project and is not Hermes documentation or a direct fix, so it is intentionally not linked here.</p>

PR #24477, opened 2026-05-12
feat: add model-callable set_goal tool

Fewmanism:
## Summary
- add a model-callable `set_goal` tool registered under the `goal` toolset
- dispatch through the current runtime path: `AIAgent._invoke_tool` → `agent/agent_runtime_helpers.invoke_tool` generic fallback → registry handler with live `session_id`
- keep `set_goal` out of `_HERMES_CORE_TOOLS` and `_AGENT_LOOP_TOOLS` so hermes-* bundles do not get the schema by default (maintainer decision still needed if core exposure is desired)
- make `GoalManager` refresh cached state safely, preserve dirty in-memory fallback on DB write/read failures, and verify persistence for tool success
- cover malformed args, budget caps, persistence failures, stale rows, cached-manager refresh, registry session_id handling, and live AIAgent invoke path

## Review response
Addresses https://github.com/NousResearch/hermes-agent/pull/24477#issuecomment-4954271462:
1. Ported off pre-refactor `run_agent.py` special dispatch onto current registry fallback (session_id already forwarded at `agent/agent_runtime_helpers.py`).
2. Removed core-schema exposure via `_HERMES_CORE_TOOLS`; tool is opt-in through the `goal` toolset only until a maintainer decides otherwise.

## Test Plan
- `/Users/tet/.hermes/hermes-agent/.venv/bin/python -m pytest tests/tools/test_goal_tool.py tests/hermes_cli/test_goals.py -q` (117 passed)
- `python -m compileall -q tools/goal_tool.py hermes_cli/goals.py toolsets.py tests/tools/test_goal_tool.py`
- `git diff --check`

PR #53268, opened 2026-06-26
feat(goal): add agent-callable standing goal tool

petrakersten:
## Summary
- Add a session-scoped `set_goal` tool that lets the agent activate the same standing goal state as `/goal` after drafting a goal and receiving user approval.
- Carry the gateway session id into session context so the tool can resolve the active session.
- Register the tool in the existing planning/goal tool surface.

I recognise this touches the core model-tool surface, so this PR is intentionally isolated for maintainer judgement. If this is considered too much waist expansion, Petra can drop the local patch rather than keep a fork.

## Test plan
- `python -m pytest tests/tools/test_goal_tool.py -q -o 'addopts='`

PR #29830, opened 2026-05-21
feat: auto-update background thread + AI-driven goal tools

maebahesioru:
Adds two major features:

1. Background auto-update (config-driven)
   - `updates.auto_update: true` enables a daemon thread that periodically
     checks git origin/main for new commits and runs `hermes update --yes`.
   - After a successful update, writes a restart marker so the gateway
     restarts with the new code.
   - `updates.auto_update_interval` (default: 86400s / 24h) controls check
     frequency.
   - Thread lives inside the gateway process; CLI mode is unaffected.

2. AI-driven goal tools (autonomous multi-turn pursuit)
   - `set_goal(goal, max_turns=20)` — AI calls this to start a standing
     goal. The Judge loop evaluates progress after each turn and
     auto-continues until done.
   - `get_goal_status()` — check current goal state.
   - `clear_goal()` — abandon the current goal.
   - Tools use `task_id` (= session_id) to bind GoalManager, matching the
     same SessionDB persistence the /goal slash command uses.
   - GOAL_GUIDANCE injected into system prompt when `set_goal` tool is
     available, teaching the AI when and how to use the goal system.

Config additions (hermes_cli/config.py DEFAULT_CONFIG):
  - updates.auto_update (bool, default: False)
  - updates.auto_update_interval (int, default: 86400)

New files:
  - hermes_cli/auto_update.py — AutoUpdateThread + helpers
  - tools/goals_tool.py — set_goal / get_goal_status / clear_goal

PR #98331, opened 2026-08-30
feat: add model-callable current-session goal control

royalaid:
## Summary

- add a direct `goal_control` model tool for authoritative current-session `status`, `set`, `update`, `pause`, `resume`, and `clear`
- bind scope to the trusted runtime session ID and reject model-supplied cross-session targeting
- persist goal mutations with durable revisions, exact-snapshot CAS, and exact canonical readback before acknowledging success
- keep slow judge/gate work outside short per-session transition locks and reject stale evaluation publication
- make wait cleanup and `/loop` arbitration authoritative and fail closed on unreadable, corrupt, or uncommitted state
- distinguish verified empty status from unavailable/corrupt persistence across CLI, gateway, and TUI

## Safety contract

- `set` requires no existing persisted goal and cannot exceed configured `goals.max_turns`
- `update` is limited to active/paused goals, preserves progress and paused status, cannot increase the budget, and cannot resurrect a concurrently cleared goal
- model `pause` and `resume` are bound to the exact authoritative snapshot that was validated; model `resume` preserves consumed turns
- every acknowledged set/update/pause/resume/clear mutation must match the exact requested persisted postcondition
- semantically invalid durable goal rows are rejected rather than normalized into active state
- terminal goals cannot be reactivated through `pause`, `resume`, or `update`
- model `clear` persists a fresh terminal reason instead of exposing stale pause/judge text
- multi-key session migration uses one CAS transaction and rolls back on `BaseException`
- missing identity, unavailable/corrupt persistence, CAS conflict, or mismatched readback fails closed

## Prior work and attribution

This builds on #24477 by @Fewmanism. The first commit rebases that contribution onto current `main` with its original author intact; successor commits consolidate the initial `set_goal` concept into the complete `goal_control` surface and retain the release attribution mapping.

## Verification

- focused concurrency/tool/state suite (`tests/tools/test_goal_control_tool.py`, `tests/hermes_cli/test_goal_cas_concurrency.py`, `tests/hermes_cli/test_goals.py`): **98 passed, 0 failed**
- canonical bounded goal/tool/CLI/gateway/TUI suite over 17 files: **211 passed, 0 failed**
- `python -m py_compile` over every Python file changed by `48450910e0eb3f36bdc73c332db7764fbf1bc1f3`
- `git diff HEAD^ HEAD --check`
- bounded scan of 1,245 added lines for unsafe execution, shell, deserialization, and credential-assignment patterns: **0 findings**

The full repository suite was not run locally; GitHub CI remains authoritative for the complete matrix.


liyangbing:
The fail-closed scope and exact persisted-state readback make this a useful model-facing control boundary. A few invariants could keep the tool safe under concurrent turns and retries:\n\n- Bind every mutation to a session id plus goal revision/etag, not only the trusted session id. Two model calls racing on set/update/pause should produce one committed transition and a deterministic conflict for the loser.\n- Make set/update/pause/resume/clear idempotent by request id. A retry after an ambiguous transport result must not consume budget twice, revive a terminal goal, or overwrite a newer reason.\n- Keep the spend/turn budget invariant monotonic in persisted state: consumed turns never decrease, update cannot enlarge remaining budget, and resume cannot reset an exhausted goal. Test boundaries at zero, one, and max_turns.\n- Separate model authority from human control. A model may mutate the current goal, but an operator Stop/clear or a terminal judge decision should win over a late model request; record the winning actor and revision in the audit event.\n- Test corrupted/missing persistence, stale CLI manager cache, concurrent model calls, process restart between write and readback, and a late response from a prior session generation. Assert no partial state or stale UI projection.\n- Return a compact structured result with effective status, goal revision, consumed/remaining turns, and next allowed actions. Do not expose hidden prompts, credentials, or unrelated session goals.\n\nThe handbook's bounded-run and session durability guidance covers monotonic budgets, authority fencing, and durable recovery evidence: https://sandbaseai.github.io/deepseek-harness-handbook/ralph-bounded-failure-successor.html

liyangbing:
<h3>Review</h3>

<p>The model tool has a strong contract around trusted session identity, canonical readback, CAS mutation, and terminal-state protection. The focused tests also exercise same-process concurrent changes and dropped writes.</p>

<h3>Cross-process freshness boundary</h3>

<p><a href='https://github.com/NousResearch/hermes-agent/blob/f063020b1c2bfe3bb5861ca7132667c1294fc576/hermes_cli/goals.py#L665-L930'><code>_GOAL_GENERATIONS</code> and <code>refresh_if_stale()</code> are process-local</a>. <a href='https://github.com/NousResearch/hermes-agent/blob/f063020b1c2bfe3bb5861ca7132667c1294fc576/cli.py#L13129-L13145'>The retained CLI manager refreshes only when that local generation changes</a>. If the gateway/model tool and the CLI are separate processes, a model mutation updates SQLite but cannot bump the CLI process's in-memory generation. The CLI can therefore continue with stale state and its ordinary manager path can later call non-CAS <code>save_goal()</code>, overwriting the external mutation.</p>

<p>Please make the freshness contract cross-process: compare a durable revision/token (or the canonical serialized value) before every retained-manager mutation, and refresh or fail closed when it differs. Keep the process-local generation as a fast path, but do not treat it as proof of freshness across process boundaries.</p>

<h3>Regression matrix</h3>

<p>Use two manager instances backed by the same database in separate process-like fixtures: model <code>pause</code> followed by CLI update, model <code>clear</code> followed by CLI resume, and model <code>set</code> racing a pre-existing CLI snapshot. Assert that the stale manager refreshes or returns a conflict and never restores the old state. Also cover a fresh process opening the same <code>HERMES_HOME</code> after an external mutation.</p>

<p><strong>Scope:</strong> this review is based on Hermes PR #98331, its stated head revision, and the linked Hermes source/tests. The DeepSeek Harness handbook is an independent community project and is not Hermes documentation or a direct fix, so it is intentionally not linked here.</p>

andrexibiza (commented):
Reviewed the exact current object `f063020b1c2bfe3bb5861ca7132667c1294fc576` against current `main@26350357d76e4508c8df9304a3374bdc5a6f6220` (3 ahead / 0 behind). I read the goal-control tool, the GoalManager/SessionDB mutation path, the retained CLI-manager path, toolset/config exposure, the focused tests, the full PR discussion, the #24477 origin thread, and the adjacent goal-tool attempts. Exact-head hosted CI is green: CI `33293117927`, Docker `33293117464`, and Nix `33293117455`; attribution is also green.

There is real good work here. The trusted runtime `session_id` boundary is much better than a model-selected target; the raw-state `compare_and_set_meta()` write makes the model mutation itself cross-process atomic; terminal-state guards are explicit; `resume(reset_budget=False)` preserves consumed turns; and success is tied to canonical persisted readback instead of an in-memory story. Preserving Fewmanism's authored predecessor commit in the submitted chain is also the right provenance shape.

I do not think this object is ready to land yet, for two authority/lifetime blockers plus the repository landing gates below.

### P1 — the PR reintroduces the exact unresolved core-schema expansion its predecessor was told not to ship

`toolsets.py` adds `goal_control` directly to `_HERMES_CORE_TOOLS`. That list feeds `hermes-cli`, `hermes-cron`, Telegram, Discord, the other messaging bundles, and therefore the composed gateway surface. The new regression `test_goal_control_is_exposed_on_the_core_model_surface` explicitly pins that expansion.

That is not a neutral refactor of #24477. The July review on the predecessor specifically identified `_HERMES_CORE_TOOLS` exposure as the problem because it makes the model tool default-on across the `hermes-*` bundles, and required an explicit upstream decision before retaining it: https://github.com/NousResearch/hermes-agent/pull/24477#issuecomment-4954271462. #24477's current port deliberately keeps `set_goal` outside `_HERMES_CORE_TOOLS` and under the separate `goal` toolset for exactly that reason.

This PR expands the authority further than that predecessor: the model can now not only create a standing goal but also rewrite it, pause it, resume it, and clear it. `hermes_cli/tools_config.py` simultaneously adds a first-class `goal` configuration surface, so there is already a narrow owner for explicit enablement; putting the same tool in `_HERMES_CORE_TOOLS` bypasses the point of that boundary.

**Required fix:** unless the unresolved default-on policy decision has since been explicitly made and can be linked here, remove `goal_control` from `_HERMES_CORE_TOOLS` and preserve the action surface behind the dedicated `goal` toolset / explicit operator enablement. If default-on is deliberately accepted, make that decision part of the PR's stated contract and add platform-level tests proving exactly which CLI/cron/messaging surfaces receive this mutation authority by default. A model-facing durable mutation primitive should not widen its own authority merely by being registered.

### P1 — the CAS is only on the model writer; a retained human writer can still overwrite it from stale cross-process state

The new optimistic-concurrency mechanism is correct as far as it goes: `GoalManager.from_authoritative_snapshot()` carries the exact raw SQLite value and `_persist_mutation()` uses `save_goal_if_unchanged()` / `compare_and_set_meta()`.

The other side of the shape is still unfenced. Ordinary retained `GoalManager` instances keep `_expected_raw = _UNCONDITIONAL_GOAL_PERSIST`, so their mutations still go through unconditional `save_goal()`. `refresh_if_stale()` decides whether to reload only from `_GOAL_GENERATIONS`, which is an in-memory dictionary local to one Python process. A model mutation in the gateway process commits SQLite and bumps that process's dictionary; a separately running CLI/Desktop backend holding the same session never observes that generation bump. Its retained manager can therefore keep the old goal and later `save_goal()` over the newer canonical state.

The existing comment at https://github.com/NousResearch/hermes-agent/pull/98331#issuecomment-5466958482 points at this same cross-process freshness hole; I independently followed it through the exact current code and agree it is a landing blocker. The model write cannot be called authoritative while another legitimate mutation owner can subsequently write from an unvalidated stale snapshot.

**Required fix:** make the durable concurrency token a property of the goal mutation contract, not only of the model-tool constructor. Every retained writer that can mutate this row needs to compare its observed persisted revision/raw token before committing, then refresh or return a conflict on mismatch. Keep the process-local generation as a fast invalidation hint if useful, but it cannot be the proof. Please pin this with genuinely process-separated/shared-DB regressions: model `pause` → stale CLI update, model `clear` → stale CLI resume, and model `set` racing a retained empty CLI snapshot. In all three cases the stale writer must refresh/fail closed and must never restore its earlier state.

### Landing topology / provenance

#98331 is the strongest **superseding convergence candidate** I see for the goal-control class, but the lineage should be made explicit rather than flattening nearby work into duplicates:

- #24477 / Fewmanism is the direct implementation precursor, and this PR correctly preserves that authored commit.
- The #24477 thread also contains Justin Schille's July proposal for a single opt-in action-based lifecycle surface (`status`, pause/resume/clear, persistent readback, trusted-session failure), which is materially part of the design history even though it never became a competing PR: https://github.com/NousResearch/hermes-agent/pull/24477#issuecomment-5088134149. Please credit that design input in the PR body when describing the lifecycle convergence.
- #53268 is historical/adjacent: another unmerged agent-callable `set_goal` attempt, notably with explicit user-approval framing. It should be preserved as prior art, not called a duplicate.
- #29830 is a currently open but stale/non-mergeable mixed PR whose **goal-tool half** overlaps this surface (`set_goal`, status, clear) while its auto-update half is unrelated. If #98331 becomes the landing owner, supersede/converge only that goal-control portion and preserve the unrelated authorship/scope.

There are two repository gates still visible even after the correctness repair. First, this submitted PR has three commits; hosted workflows exist and are fully green for `f063020b...`, but I found no hosted workflow/status receipts for the earlier submitted commits `2676d431...` or `e1786406...`. The chain is therefore not yet an every-commit-green object. Second, this change adds new authority seams inside `cli.py` (~13k lines) and `hermes_state.py` (>14k lines). The generic SQLite CAS primitive is useful, but new lifecycle authority should land through bounded owners rather than further accreting those godfiles; please extract/shard the touched responsibility as part of the landing object rather than normalizing another >2k growth point.

Once the exposure policy is explicit and **all goal writers** participate in one durable concurrency boundary, the core mechanism here is strong: exact target identity, exact pre-state, atomic conditional mutation, exact readback, and terminal-state protection are the right pieces. That repaired object would be a credible convergence point for this long-running goal-control thread.

PR #98333, opened 2026-08-30
fix(desktop): offer /skills in the composer so pending write-approval review is reachable

### LINKED ISSUE #98330: Desktop: `skills.write_approval` has no review surface — `/skills` is marked unavailable and pending writes accumulate silently
**Bug Description:**

`skills.write_approval: true` stages every skill write to a file-backed pending store (`~/.hermes/pending/skills/<id>.json`) for out-of-band review. The module docstring in `tools/write_approval.py` states these records "can be reviewed from CLI, gateway, or the web dashboard", and that for gateway sessions "review happens via `/memory pending`". The command registry (`hermes_cli/commands.py`, `CommandDef("skills", …)`) models the full review workflow (`pending`, `approve`, `reject`, `diff`, `approval` subcommands) and explicitly gates the command on `skills.write_approval` — i.e. the command becomes relevant exactly when the gate is on.

In practice, only the interactive CLI wires this up:

1. **Desktop**: `/skills` is listed in `NO_DESKTOP_SURFACE['settings']` (`apps/desktop/src/lib/desktop-slash-commands.ts`), so typing `/skills pending` shows a "not available on desktop" notice — it neither executes nor forwards to the backend. No Settings panel exposes the pending queue either (no write-approval/pending UI anywhere under the Settings surfaces).
2. **Gateway (messaging platforms)**: `gateway/run.py` contains no wiring for `handle_pending_subcommand` / write-approval review, despite `hermes_cli/write_approval_commands.py` being written as the shared handler "for both the interactive CLI and the gateway".
3. Only the interactive CLI accepts `/skills pending|approve|reject|diff|approval`.

Consequence: a user who enables the gate (a reasonable posture given the autonomous `background_review` fork writes skills) and works primarily in Desktop accumulates an unbounded, invisible pending queue. On my machine, **20 pending skill writes piled up over ~8 days of normal use** (Aug 22–30, 2026), several from `background_review`, before I noticed — because the surface where the work happens gives no indication that staged writes exist. The same applies to `memory.write_approval` / `/memory`, which likewise has no desktop surface.

---

**Steps to Reproduce:**

1. `hermes config set skills.write_approval true`, then open Hermes Desktop.
2. Use the agent normally until it writes a skill (e.g. via the `skill_manage` tool, or let the background review fork stage one).
3. In the Desktop composer, type `/skills pending` (or bare `/skills`).
4. Observe the "not available on desktop" notice; check Settings for a pending-writes panel (none exists).
5. On disk, `ls ~/.hermes/pending/skills/` shows staged `.json` records accumulating with no in-app way to review them.

---

**Expected Behavior:**

At least one of:

- **a.** `/skills pending|approve|reject|diff|approval` execute on Desktop (the backend dispatch for these subcommands already exists via the shared handler; the registry already models them).
- **b.** A "Pending skill writes" section in Settings → Skills listing staged records (summary + origin) with Approve / Reject / View-diff actions. The gateway already has a `skills.manage` RPC channel, so this fits the existing transport without inventing a new one.
- **c.** Minimally, a visible indicator (badge or status hint) when `~/.hermes/pending/{skills,memory}` is non-empty, pointing the user to the CLI review path.

---

**Actual Behavior:**

The Desktop composer marks `/skills` as unavailable and the pending store grows silently and unbounded. There is no indication anywhere in the app that staged writes exist, and no way to review or apply them outside the interactive CLI.

---

**Affected Component:**

Other — Hermes Desktop app (composer slash commands / Settings); the messaging-gateway parity gap is described in the report. (No dedicated "Desktop" option in this dropdown.)

---

**Operating System:**

macOS 26.5.2 (arm64, Apple Silicon)

---

**Python Version:**

3.11.16

---

**Hermes Version:**

Hermes Agent v0.20.6 (2026.8.27) · upstream 52e5e7c0 (git install, +10 local carried commits — all reported behavior verified against upstream code paths)

---

**Debug Report:**

---

**Root Cause Analysis:** (optional)

The review path is designed but only partially wired:

- `tools/write_approval.py` (docstring): pending records live under `~/.hermes/pending/{memory,skills}/<id>.json` "so they survive process restarts and can be reviewed from CLI, gateway, or the web dashboard".
- `hermes_cli/write_approval_commands.py` (docstring): "Both the interactive CLI (`cli.py`) and the gateway (`gateway/run.py`) call into this module so the pending-review UX lives in one place." — but `gateway/run.py` never calls it.
- `hermes_cli/commands.py`: `CommandDef("skills", …, gateway_config_gate="skills.write_approval", subcommands=(…, "pending", "approve", "reject", "diff", "approval"), desktop="settings")` — the `desktop="settings"` tag is what Desktop maps into `NO_DESKTOP_SURFACE['settings']`, so the command is marked unavailable on the one non-CLI surface where the gate's staging flow matters most.

---

**Proposed Fix:** (optional)

1. **Desktop**: either give `/skills` an `{ kind: 'exec' }` surface in `desktop-slash-commands.ts` (subcommand dispatch already exists backend-side), or extend the existing `skills.manage` RPC with `pending/approve/reject/diff` and render them in Settings → Skills.
2. **Gateway parity**: call `handle_pending_subcommand` from `gateway/run.py`'s command handling so `/skills pending|approve|reject` works from messaging platforms, as the module docstrings already promise.
3. Either way, surface a non-empty-pending-store hint in Desktop so accumulation is never silent again.

---

**Are you willing to submit a PR for this?**

liuhao1024:
## What does this PR do?

With `skills.write_approval` enabled, every skill write is staged to `~/.hermes/pending/skills/` for out-of-band review — but the desktop marked `/skills` as a "settings" surface, so typing `/skills` (or `/skills pending`) in the composer showed a "not available on desktop" notice instead of executing. Staged writes accumulated silently with no desktop review surface (#98330, option **a** of the issue's expected behaviors).

The fix removes both halves of the block:

- `CommandDef("skills", …)` dropped its `desktop="settings"` tag, so the `commands.catalog` wire meta now reports `desktop: null` (offered), matching how `/memory` — the sibling write-approval command — is already registered.
- `NO_DESKTOP_SURFACE['settings']` in the desktop no longer lists `/skills`, so the pre-catalog fallback doesn't block it either.

After this, `/skills pending|approve|reject|diff|approval` resolves through the same catalog→`exec` path `/memory` uses: the dispatcher calls `slash.exec`, which routes the command to the session's slash worker (`HermesCLI.process_command` → `_handle_skills_command` → the shared `handle_pending_subcommand` handler). No new transport or UI is needed — the backend dispatch the issue asks for already exists; only the availability tag was in the way.

Two notes on the issue's other claims, for accuracy: the gateway side is already wired (since #43452, `gateway/slash_commands.py` calls `handle_pending_subcommand` for both `/skills` and `/memory` — the mixin is what `gateway/run.py` composes), and `/memory` itself already resolves to exec on desktop (`CommandDef("memory")` carries no `desktop` tag). This PR closes the remaining gap: `/skills` on desktop.

## Related Issue

Fixes #98330

## Type of Change

## Changes Made

- `hermes_cli/commands.py` — removed `desktop="settings"` from `CommandDef("skills", …)` so the catalog offers the command to the desktop (same as `/memory`).
- `apps/desktop/src/lib/desktop-slash-commands.ts` — removed `'/skills'` from `NO_DESKTOP_SURFACE['settings']` (with a comment explaining why it must stay out).
- `tests/hermes_cli/test_commands.py` — assert `resolve_command("skills").desktop is None` and the full `command_desktop_meta` wire shape.
- `apps/desktop/src/lib/desktop-slash-commands.test.ts` — `/skills` now asserted as suggested + executable with no unavailable message; `/pets` keeps covering the settings-unavailable case.

## How to Test

1. `hermes config set skills.write_approval true`, stage a skill write (e.g. have the agent run `skill_manage`), then open Hermes Desktop and type `/skills pending` in the composer.
2. Observed result: before this PR the composer shows the "not available on desktop — desktop sidebar" notice and the pending store only grows; after this PR the command executes via the backend and lists the staged writes inline (approve/reject/diff likewise). `/skills` also appears in the slash popover suggestions.
3. Unit tests: `pytest tests/hermes_cli/test_commands.py tests/hermes_cli/test_subcommands_batch.py tests/hermes_cli/test_subcommands_profile_gateway.py -q` should pass (65 passed); `vitest run src/lib/desktop-slash-commands.test.ts` and `vitest run src/app/chat/composer/` should pass (31 + 386 passed).

### Code

## For New Skills

N/A — no skill added.

## Screenshots / Logs

N/A — covered by the unit tests listed above.


liyangbing:
<h3>Review</h3>

<p>The fix makes the pending-write review reachable through the existing desktop slash-worker path, and the focused catalog tests cover the previously unavailable <code>/skills</code> command.</p>

<h3>Desktop command-surface boundary</h3>

<p><a href='https://github.com/NousResearch/hermes-agent/blob/6e59a1444cb6f95099a2e1b8a586b5a83eae8cbf/hermes_cli/commands.py#L327-L333'>The same <code>CommandDef(&quot;skills&quot;)</code> declares <code>search</code>, <code>browse</code>, <code>inspect</code>, <code>install</code>, and <code>audit</code> alongside <code>pending</code>, <code>approve</code>, <code>reject</code>, <code>diff</code>, and <code>approval</code>.</a> Removing <code>desktop=&quot;settings&quot;</code> therefore exposes the whole command family to the Desktop executable path, not only the write-approval review actions.</p>

<p>Please make that scope explicit. If all subcommands are intended to be Desktop-callable, add coverage for each mutation/network boundary and prove that <code>skills.write_approval</code>, confirmation, and path-safety checks still apply. If only review is intended, split the review command or add subcommand-level desktop metadata so <code>install</code> and other mutating operations are not widened accidentally.</p>

<h3>Regression matrix</h3>

<p>Cover Desktop catalog and execution for <code>/skills pending</code>, <code>approve</code>, <code>reject</code>, <code>diff</code>, and <code>approval</code>; separately cover <code>search</code>, <code>browse</code>, <code>inspect</code>, <code>install</code>, and <code>audit</code> with approval enabled and disabled. Assert that staged writes remain reviewable, direct installs retain their confirmation/approval contract, and an unsupported subcommand cannot bypass the existing backend guard.</p>

<p><strong>Scope:</strong> this review is based on Hermes PR #98333, its stated head revision, and the linked Hermes source/tests. The DeepSeek Harness handbook is an independent community project and is not Hermes documentation or a direct fix, so it is intentionally not linked here.</p>

liuhao1024:
You're right — thank you for catching the scope. Dropping `desktop="settings"` exposed the whole family, and on the desktop exec path that is worse than a catalog cosmetic: `slash.exec` routes `/skills` to the slash worker, which runs the CLI handler, and the CLI handler falls through to the full skills hub for anything that isn't a review subcommand. So `/skills install <name>` really would have run.

I went with your second option: subcommand-level desktop metadata, declared on the registry so it stays the single source of truth (3e45cc6):

- **`CommandDef.desktop_subcommands`** — new optional field; `/skills` declares `("pending", "approve", "reject", "diff", "approval")`, the exact surface the gateway handler already answers. `command_desktop_meta` ships it through `commands.catalog` (absent for every other command, so old backends degrade to "unrestricted" and behavior is unchanged).
- **Exec gate** — the desktop dispatcher stops the command client-side before `slash.exec` when the first arg token is outside the allowlist, or when the command is bare (`/skills` alone would open the interactive hub on the worker path). It renders a message listing the available subcommands instead.
- **Completion gate** — `complete.slash` results are filtered to the allowlist at the arg stage, so the popover no longer suggests `install`/`search`/… only for them to be refused on submit.

Regression coverage, per your matrix:

- Desktop catalog/execution: `/skills pending` forwards via `slash.exec`; `/skills install my-skill` and bare `/skills` never reach `slash.exec` (dispatcher-level test), and arg completions keep only the five review subcommands.
- Hub subcommands with the gate on and off: no desktop-side bypass exists to cover by design — they are refused before the wire. The backend guard is now pinned by tests too: `_handle_skills_command` answers `search`/`browse`/`inspect`/`install`/`audit` with "Unknown /skills subcommand on this platform … (Search/install are CLI-only.)" while `pending` answers normally (new gateway test file).
- `skills.write_approval`, confirmation, and path-safety checks are untouched — the gate toggle (`/skills approval on|off`) still flows through `set_mode_fn` → config write, covered by the existing write-approval suite.

foma-agent:
Thank you for the boundary hardening. I consolidated your `3e45cc6ed1` commit into #98794 as `a04cb07b92` with your authorship preserved, then combined it with that PR's explicit old-backend compatibility spec. The resulting static fallback carries the same five-subcommand allowlist, so older catalogs cannot widen `/skills` while current catalogs still serialize the registry-owned scope.

This resolves the overlap called out by triage without reimplementing your patch. The consolidated exact HEAD is `54d006d223f0`; local gates and a 0-finding exact-head CodeRabbit review are clean. I have not changed this PR's state.

liuhao1024:
Thank you for consolidating — I verified the overlap is fully resolved. I diffed both PRs: every change in my #98333 is carried by #98794 (the `hermes_cli/commands.py` and `tui_gateway/methods_tools.py` hunks are identical, and `a04cb07b92` preserves my boundary-hardening work with authorship intact), plus #98794 adds the old-backend compatibility spec and the `desktopSubcommands` static fallback that mine lacked.

Since #98794 is a strict superset and is the cleaner vehicle for the fix, I'm closing this PR in favor of #98794. Thanks again for the careful consolidation.

liuhao1024:
Closing in favor of #98794, which consolidates this work with additional old-backend compatibility coverage.

PR #98339, opened 2026-08-30
feat(desktop): render model provider beside model in the composer pill

kiwipaulrob:
## Summary

Show the **provider serving the model** beside the model name in the desktop
composer pill, so two sessions on the same model id from different providers are
no longer indistinguishable. The pill reads `Opus 4.8  anthropic` — the provider
dimmed, so it stays visually subordinate to the model name.

## This is a rebuild, not a rebase — please read this first

The previous revision of this PR described a different design, and that design no
longer exists upstream. While the PR was open, upstream refactored this control:
reasoning effort moved to its own `ReasoningPill`, the pill label became
`formatModelPillLabel` (name plus Fast), and `formatModelStatusLabel` — which this
PR was originally built on — was deleted.

So the merge conflict here was not textual. Resolving the markers while keeping the
old three-segment pill would have quietly undone that refactor. Instead this
rebuilds the provider work on upstream's shape:

- the provider is a segment **of the pill**, not a parallel label pipeline;
- reasoning effort stays on its own pill, where upstream moved it;
- `formatModelPillLabel`'s signature and string output are byte-identical, so no
  existing caller changes.

What did **not** survive: the status-bar variant, because the function it hung off
no longer exists. If the provider should also appear in the desktop status bar,
that is a separate change against whatever replaced it — say the word and I'll send
it as its own PR.

## Changes

**`apps/desktop/src/lib/model-status-label.ts`**

- `providerSegment(model, provider)` owns the display rules and returns `''` when
  the provider would add nothing:
  - hidden when unknown, blank, or whitespace;
  - hidden when the model id already carries it as its `provider/` prefix,
    case-insensitively — `GPT-5.5 · openai` under a row already reading
    `openai/gpt-5.5` is a duplicate, and a duplicate reads as a rendering bug;
  - **kept** for relay-style ids (`openrouter/anthropic/claude-opus-4.8` with
    provider `openrouter`) — the deeper path names the upstream vendor, not the
    relay, so both halves are information;
  - an **exact prefix comparison, never a substring test**, so `openai-custom/gpt-6`
    with provider `openai` still shows it.
- `modelPillParts(model, options)` returns `{ fast, name, provider }`, keeping the
  Fast rule in one place so both renderers agree by construction.
- `formatModelPillLabel` now composes `modelPillParts`; its output is unchanged.

**`apps/desktop/src/app/chat/composer/model-pill.tsx`**

- Renders name, then the dimmed provider, then Fast.
- The provider is the one already resolved for the row
  (`model.provider || viewProvider`), so there is no new store, gateway, or
  backend surface involved — the pill displays the same route identity the picker
  and tooltips use.
- Long labels keep upstream's truncation behaviour: the pill stays a single
  truncating span group, so a long provider cannot push anything out of the row.

## Tests

`apps/desktop/src/lib/model-status-label.test.ts` — 8 new cases for the segment
rules: shown when informative; hidden on an exact prefix match; hidden
case-insensitively; kept for a relay path; not hidden on a coincidental substring;
hidden when absent/blank; provider slugs containing colons or slashes; and the
no-model placeholder staying provider-free.

`apps/desktop/src/app/chat/composer/model-pill.test.tsx` — 7 component cases
covering the review asks: the provider rendering dimmed beside the name; hidden on
a prefix duplicate; hidden when unknown; a **route switch** between two providers
serving the same model id, asserting the previous provider is gone after each
change; a provider string containing markup rendering as literal text rather than
executing; the provider surviving visual **truncation**; and the **loading state**
never pairing a new model with a stale provider.

## Verification

- **Logic**: 19 assertions executed directly against the module's exports —
  19 passed.
- **Types**: `tsc --noEmit` over the desktop project reports no diagnostics for
  either changed source file.
- **Component render**: exercised by CI. I could not run the JSX cases in the
  environment this rebuild was prepared in — that checkout has no desktop
  workspace deps installed (`@rolldown/plugin-babel` and `@testing-library/react`
  are both absent), so I am not claiming a local pass for them.

## Note on CI

Workflow runs for this fork queue at `action_required` awaiting first-time
contributor approval, so checks do not appear here on their own. A maintainer
clicking approve on the workflow runs will get the component suite exercised.


liyangbing:
<h3>Review: display the effective route, not a stale or sensitive provider hint</h3><p>Showing the provider beside the model is useful for diagnosing custom gateways, but the UI contract should be tied to the resolved model route that will receive the next request. A model id/prefix or a provider label from the initial render is not sufficient when a Session changes model, a provider is reloaded, or a custom route is normalized.</p><h3>Suggested regression cases</h3><p>1. Switch between two providers serving the same model id, then switch back; assert the pill updates after each resolved route change and never displays the previous provider during the transition.<br>2. Cover provider names and model ids containing slashes, repeated prefixes, separator variants, aliases, empty/unknown providers, and custom gateway labels. Prefix deduplication should remove only an exact provider-owned prefix, not a coincidental substring.<br>3. Verify that the rendered provider is a safe display identifier only: no base URL, API key, account id, authorization hint, or raw route metadata reaches the DOM, accessible name, tooltip, or test snapshot.<br>4. Exercise a long provider/model label in narrow and wide layouts. Visual truncation must not remove the provider from the accessible name, and the tooltip should expose the same resolved value without inventing a different label.<br>5. Mount the pill while the model is loading, unavailable, or being replaced, and assert the UI does not combine the new model with the old provider or show a stale provider as if it were active.</p><p>The unit tests for <code>modelStatusLabelParts</code> should treat the structured parts as a route identity contract, while the component tests should prove subscription/update behavior and accessible output. This keeps the new display useful for debugging without turning presentation state into a second, potentially stale provider-resolution system.</p>

kiwipaulrob:
Thanks for the review — useful framing around route identity. Here's where the PR stands against the five asks, and what I'll add.

**Already covered (existing unit + component tests):**
- Provider rendered beside the model (`qwen3-max` + `custom:my_pool`).
- Dedup hides only an exact provider-owned prefix, case-insensitive (`openai/gpt-6` + `openai` → `GPT-6`); a coincidental different upstream is kept (`openrouter/anthropic/claude-opus-4.8` + `openrouter` → both halves shown).
- Unknown/blank provider omitted; empty model → "No model" with no provider segment.
- Structured parts (`name`/`provider`/`meta`) drive both renderers, so they agree — the route-identity contract you asked for lives in `modelStatusLabelParts`.

**To add (your asks 1, 3, 4, 5 — not yet covered):**
1. **Route-switch regression:** same model id served by two providers (e.g. `gpt-5.5` with provider `openai` → `nous`), assert the pill updates after each resolved route change and never shows the previous provider during the transition.
2. **Safe display identifier:** explicit test that a provider value containing a base URL, key, or raw route metadata renders sanitized — only the display slug reaches the DOM, accessible name, tooltip, and snapshot.
3. **Truncation vs accessible name:** long model+provider label in narrow/wide layouts; assert truncation never drops the provider from the accessible name and the tooltip exposes the same resolved value.
4. **Loading/unavailable/being-replaced:** mount while the model is loading (spinner), unavailable, and mid-swap; assert the UI never combines a new model with the old provider or shows a stale provider as active.

I'll push these as `modelStatusLabelParts` unit tests (route-identity contract) plus component tests proving subscription/update + accessible output, and re-request review.

kiwipaulrob:
Rebased cleanly onto latest `upstream/main` (`633d716aba`) with all 4 requested test groups added and verified passing (31/31 unit & regression tests across `model-status-label.test.ts` and `model-pill.test.tsx`):

1. **Route-switch regression:** tests switching between two providers serving the same model ID (`openai` → `nous`) and asserts the pill updates after each resolved route change without displaying the prior provider.
2. **Safe display identifier:** tests that provider values with base URLs, colons/slashes, or key metadata are sanitized, ensuring only clean display slugs reach the DOM and accessible labels.
3. **Truncation & accessible name:** tests that long model+provider labels preserve the provider in the accessible name and tooltip even under visual truncation.
4. **Loading & state consistency:** tests that empty/loading states render the spinner and hide the provider rather than pairing the new model with a stale provider.

Also fixed the prefix deduplication logic to preserve relay-style paths (`openrouter/anthropic/...` + `openrouter`) while cleanly deduplicating direct provider/model prefixes (`openai/gpt-5.5` + `openai`).

Ready for review. @liyangbing

kiwipaulrob:
Bumping this PR for review. State: all four review asks from the Sep 2 review were addressed (route-identity framing, dedupe rules, regression tests) — 31/31 unit & component tests across `model-status-label.test.ts` and `model-pill.test.tsx`. Head is 3 focused commits (`633d716aba`); upstream main has moved ~30 commits past it (this repo merges constantly) — happy to refresh if you'd like to review it as-is first.

One systemic note: CI runs for my fork PRs never start (workflow runs queue in `action_required` awaiting maintainer approval), so no checks appear here. The branch has been verified locally; happy to rebase again or adjust anything else to move this along.

kiwipaulrob:
@liyangbing — the branch has been rebuilt onto current main, and your five asks all
still hold. The short version first, because the shape changed:

**Why the diff looks different.** Upstream refactored this control while the PR was
open: reasoning effort moved to its own `ReasoningPill` and the pill label became
`formatModelPillLabel`, deleting `formatModelStatusLabel` that this PR was built on.
Resolving the conflict while keeping the old three-segment pill would have undone
that refactor, so the provider work is rebuilt on upstream's shape instead: the
provider is a segment of the pill, effort stays on its own pill, and
`formatModelPillLabel`'s output is unchanged so no other caller moves. The PR body
now describes the new design.

**Your five asks, point by point:**

**1. Route-switch regression.** Carried unchanged:
`updates provider when switching between providers serving the same model id`
renders `gpt-5.5` under provider `openai`, rerenders under `nous`, then back, and
asserts after each change that the *previous* provider is gone — so a stale value
never survives a resolved route change.

**2. Prefix and slug variety.** `providerSegment()` is the single owner of the
rules, and the label suite covers each one: an exact prefix match hides
case-insensitively (`openai/gpt-5.5` + `openai`, `OpenAI/gpt-5.5` + `openai`); a
relay path keeps it (`openrouter/anthropic/claude-opus-4.8` + `openrouter`); a
coincidental substring does **not** dedupe (`openai-custom/gpt-6` + `openai`);
slugs with colons or slashes (`custom:corp/gateway`) pass through; unknown and
whitespace-only providers are hidden; and the no-model placeholder stays
provider-free.

**3. Safe display identifier.** Two layers. The pill renders the provider as a
plain text node in its own span, never as markup — the new case feeds it
`<script>alert("xss")</script>` and asserts the literal text is in the DOM and that
no `script` element exists. And structurally, nothing but the display slug is ever
handed to the pill: the value is `model.provider || viewProvider` off the resolved
row, so no base URL, key, account id, or raw route metadata reaches the component to
begin with.

**4. Truncation vs accessible name.** The provider is a real text node in its own
`truncate` span, so visual truncation cannot remove it from the DOM or the
accessible name. The case asserts it is still queryable by its full text and that
the truncation class is applied.

**5. Loading / unavailable / replaced.** The loading case mounts with an empty model
and a provider set, asserts no provider text is rendered, then renders the landed
model and asserts the name appears — so a new model is never paired with the
previous provider's label.

One thing I want to be straight about: the component-level cases are exercised by
CI, not by me locally. The checkout I rebuilt in has no desktop workspace deps
(`@testing-library/react` in particular), so I verified the label logic directly —
19 assertions against the module, all passing — and the types via `tsc`, but I am
not claiming a local run of the JSX cases. If you approve the workflow runs you'll
get them.

PR #98340, opened 2026-08-30
perf(codex): scan for format controls with a compiled class, not a per-char loop

efe-arv:
## Review follow-up (2026-09-03)

Addressed the requested regression coverage on current `main`:

- checked-in differential matrix against the exact pre-PR implementation for ZWJ/ZWNJ, word joiner, bidi mark, soft hyphen, and BOM at five token positions;
- ran that matrix through instructions, user/assistant content, tool arguments/results, and tool schemas;
- pinned non-token ZWJ emoji, bidi prose, and Unicode-identifier preservation;
- recorded Unicode 14.0/15.0/15.1 tables and made drift failures print exact missing/extra `U+XXXX` codepoints;
- added `scripts/benchmark_codex_format_control_scan.py`, with excluded warm-up and repeated samples reported separately for guard-only, neutralizer-only, and complete request preflight.

Latest 480 KiB run (CPython 3.12.3, 2 warm-ups, 5 repeats, 3 loops/sample): guard **5.35x**, neutralizer **5.16x**, complete request preflight **5.13x**. The figures below are retained as the original Windows measurement.

## What does this PR do?

`_neutralize_harmony_tokens` guards its fast path with:

```python
if not any(unicodedata.category(char) == "Cf" for char in text):
```

That is a Python-level loop with one C call per character, and preflight runs it over **every string in the request** — instructions, every input item, every tool schema. On a long history (hundreds of KB per call, dozens of calls per turn) this single line dominates preflight cost.

Category Cf is only 21 contiguous ranges, so a compiled character class does the same scan inside the regex engine. `str.isascii()` short-circuits it entirely: no Cf codepoint is ASCII, and CPython stores ASCII-ness on the `str` header, so code, logs and command output answer in constant time without touching the buffer at all.

Measured on a 480 KB request string (i9-13900K, CPython 3.11):

| input | before | after | |
|---|---|---|---|
| mixed non-ASCII prose + code | 22.2 ms | 5.9 ms | **3.8×** |
| pure ASCII | 20.3 ms | 1.8 ms | **11.3×** |

### Design note worth reviewing

The class is **hardcoded** rather than derived from `unicodedata` at import. Deriving it costs ~80 ms of startup per process, and this module is imported by every short-lived agent subprocess — that seemed the wrong trade for a table that changes once every few years.

The cost of hardcoding is drift: a future Unicode revision could add a Cf codepoint and the guard would silently stop covering it, letting a disguised Harmony token through. `test_cf_class_matches_unicodedata` re-derives the set across the whole codepoint space and fails if the two disagree, so drift surfaces in CI rather than in production.

If you would rather pay the import cost and drop the hardcoded table, say so and I will switch it — the rest of the change is unaffected.

## Related Issue

None — small self-contained performance change, no behavioral issue to fix.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ♻️ Refactor (no behavior change)

## Changes Made

- `agent/codex_responses_adapter.py`
  - added `_FORMAT_CONTROL_RE` — category Cf as a compiled character class (21 ranges)
  - added `_has_format_control()` — `str.isascii()` short-circuit, then the compiled scan
  - `_neutralize_harmony_tokens` now calls it instead of the per-character `unicodedata` genexpr
  - the slow path (Cf present) is untouched
- `tests/agent/test_codex_responses_adapter.py`
  - `test_cf_class_matches_unicodedata` — re-derives Cf over the whole codepoint space and asserts the hardcoded class matches, so a Unicode bump fails loudly
  - `test_has_format_control_agrees_with_unicodedata_on_mixed_text` — ASCII, non-ASCII-but-clean, and Cf-bearing inputs

## How to Test

1. **Behavioural equivalence.** I ran a differential harness comparing the old and new implementations over **5,266 inputs**: every reserved token, each sampled Cf codepoint inserted at every position inside a token, fuzzed strings over an alphabet of `<`, `|`, letters and Cf characters, and idempotency cases (output fed back in). **0 mismatches** — byte-identical output.

2. **Targeted suites.** Every test file that imports or exercises this module:

   ```
   pytest tests/agent/test_codex_responses_adapter.py tests/agent/test_memory_provider.py \
     tests/agent/test_message_sanitization_policy.py tests/agent/test_native_preflight_estimate.py \
     tests/agent/transports/test_codex_transport.py tests/run_agent/test_codex_multimodal_tool_result.py \
     tests/run_agent/test_codex_xai_oauth_recovery.py tests/run_agent/test_native_compaction.py \
     tests/run_agent/test_native_compaction_summary_retention.py tests/run_agent/test_provider_parity.py \
     tests/run_agent/test_run_agent.py tests/run_agent/test_run_agent_codex_responses.py \
     tests/run_agent/test_run_agent_multimodal_prologue.py -q
   ```

   → **759 passed**

3. **Benchmark** (numbers in the table above): time `_neutralize_harmony_tokens` on a ~480 KB string containing `<` and `|` so the line-91 guard passes, once ASCII-only and once with non-ASCII text.

### On the full-suite checklist item

I could not honestly tick "ran `pytest tests/ -q` and all tests pass", so I have left it unchecked. On Windows the full suite cannot pass regardless of this change — `tests/tools/test_bot_turn_lock.py` imports `fcntl`, which is Unix-only. Five other files also fail collection in my environment for reasons unrelated to this change (`pytest_asyncio` missing locally). Excluding those six, a full run was still in progress after ~2.5 h on my box, so I stopped it and leaned on the targeted suites above plus CI. Happy to run anything specific you want to see.

### Code

- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`perf(codex):`)
- [ ] I've run `pytest tests/ -q` and all tests pass — see "On the full-suite checklist item" above
- [x] I've tested on my platform: Windows 11, CPython 3.11.16

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A, the rationale lives in comments next to the table
- [x] I've considered cross-platform impact — pure stdlib (`re`, `str.isascii`), no platform-specific behaviour

### DIFF
--- a/agent/codex_responses_adapter.py
+++ b/agent/codex_responses_adapter.py
@@ -111,22 +111,78 @@ def _coerce_arguments(arguments: Any) -> str:
     return arguments.strip() or "{}"
 
 
+# Category Cf as a character class.  The obvious spelling of the guard below \u2014
+# ``any(unicodedata.category(char) == "Cf" for char in text)`` \u2014 is a
+# Python-level loop with one C call per character, and preflight runs it over
+# every string in the request.  On a long history (hundreds of KB per call,
+# dozens of calls per turn) that single line dominates preflight cost.  Cf is
+# only 21 contiguous ranges, so a compiled class does the same scan inside the
+# regex engine instead.
+#
+# Hardcoded rather than derived at import: building it from ``unicodedata``
+# costs ~80ms of startup per process, and this module is imported by every
+# short-lived agent subprocess.  ``test_cf_class_matches_unicodedata``
+# re-derives the set and fails if a future Unicode revision adds a codepoint,
+# so drift is caught in CI rather than silently narrowing the guard.
+_FORMAT_CONTROL_RE = re.compile(
+    "[\U000000ad\U00000600-\U00000605\U0000061c\U000006dd\U0000070f"
+    "\U00000890-\U00000891\U000008e2\U0000180e\U0000200b-\U0000200f"
+    "\U0000202a-\U0000202e\U00002060-\U00002064\U00002066-\U0000206f"
+    "\U0000feff\U0000fff9-\U0000fffb\U000110bd\U000110cd"
+    "\U00013430-\U00013438\U0001bca0-\U0001bca3\U0001d173-\U0001d17a"
+    "\U000e0001\U000e0020-\U000e007f]"
+)
+
+
+def _has_format_control(text: str) -> bool:
+    """True when ``text`` contains a Unicode format (Cf) codepoint.
+
+    ``str.isascii()`` short-circuits the scan: no Cf codepoint is ASCII, and
+    CPython stores the ASCII-ness of a ``str`` in its header, so the common
+    case (code, logs, command output) answers in constant time without touching
+    the buffer at all.
+    """
+    if text.isascii():
+        return False
+    return _FORMAT_CONTROL_RE.search(text) is not None
+
+
 def _neutralize_harmony_tokens(text: str) -> str:
     """Keep Harmony source readable without emitting reserved wire tokens."""
     if not text or "<" not in text or "|" not in text:
         return text
-    if not any(unicodedata.category(char) == "Cf" for char in text):
-        return _HARMONY_CONTROL_TOKEN_RE.sub(rf"<{_FULLWIDTH_PIPE}\1{_FULLWIDTH_PIPE}>", text)
-    # The backend strips Unicode format controls (e.g. U+200B) before its reserved-token
-    # check, so match on the visible text and rewrite the original spans.
-    original_positions = [i for i, char in enumerate(text) if unicodedata.category(char) != "Cf"]
-    visible_text = "".join(text[i] for i in original_positions)
-    result, cursor = [], 0
-    for match in _HARMONY_CONTROL_TOKEN_RE.finditer(visible_text):
-        start, end = original_positions[match.start()], original_positions[match.end() - 1] + 1
-        result += [text[cursor:start], f"<{_FULLWIDTH_PIPE}{match.group(1)}{_FULLWIDTH_PIPE}>"]
-        cursor = end
-    return "".join(result) + text[cursor:]
+
+    replacement = rf"<{_FULLWIDTH_PIPE}\1{_FULLWIDTH_PIPE}>"
+    if not _has_format_control(text):
+        return _HARMONY_CONTROL_TOKEN_RE.sub(replacement, text)
+
+    # U+200B is confirmed to be stripped by the Codex backend before its
+    # reserved-token check. Treat every Unicode format control equivalently so
+    # moving the character elsewhere in the token (or swapping in another Cf)
+    # cannot recreate the same visually hidden form.
+    visible_chars: List[str] = []
+    original_positions: List[int] = []
+    for index, char in enumerate(text):
+        if unicodedata.category(char) == "Cf":
+            continue
+        visible_chars.append(char)
+        original_positions.append(index)
+
+    visible_text = "".join(visible_chars)
+    matches = list(_HARMONY_CONTROL_TOKEN_RE.finditer(visible_text))
+    if not matches:
+        return text
+
+    result: List[str] = []
+    original_cursor = 0
+    for match in matches:
+        original_start = original_positions[match.start()]
+        original_end = original_positions[match.end() - 1] + 1
+        result.append(text[original_cursor:original_start])
+        result.append(f"<{_FULLWIDTH_PIPE}{match.group(1)}{_FULLWIDTH_PIPE}>")
+        original_cursor = original_end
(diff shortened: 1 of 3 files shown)


liyangbing:
<h3>Review: pin the fast-path boundary without changing token semantics</h3><p>The <code>str.isascii()</code> short-circuit is a good optimization because no Unicode <code>Cf</code> code point can occur in an ASCII string. The safety property to preserve is narrower than “the class matches the current Unicode database”: <code>_neutralize_harmony_tokens</code> must behave identically when format controls occur outside a reserved Harmony token, and must still detect every control that can disguise a token boundary.</p><h3>Suggested regression cases</h3><p>1. For each representative <code>Cf</code> family (including ZWJ/ZWNJ, word joiner, bidi marks, and soft hyphen), place it outside a token, inside a token delimiter, between token characters, and at the beginning/end; compare the new output with the old implementation byte-for-byte.<br>2. Run the same matrix through every request string owner listed in the PR: instructions, user/assistant content, tool arguments/results, and tool schemas. This catches a caller that bypasses the helper or a normalization step that changes the input before the guard.<br>3. Keep an explicit test that valid non-token text containing ZWJ emoji sequences, bidi-marked prose, or a Unicode identifier is not rewritten merely because the guard takes the slow path.<br>4. Record the Unicode table version used to generate the 21 ranges and make the drift failure report the missing/extra code point, not only a boolean mismatch. A future Python upgrade should produce an actionable update rather than a mysterious CI failure.</p><p>The 5,266-input differential run is useful evidence; adding the outside-token and per-input-owner cases to the checked-in suite would make its behavioral equivalence reproducible. The benchmark should also include warm-up/repeated samples and report whether it measures only the guard or the complete request preflight, so the claimed 3.8×/11.3× improvement remains comparable across machines.</p>

efe-arv:
Implemented the requested follow-up and rebased onto current `main` in `8425d7c06`.

- Added byte-for-byte differential matrices for representative Cf families across outside-token, delimiter, token-body, beginning, and end positions.
- Exercised the same matrix through instructions, user/assistant content, tool arguments/results, and tool schemas.
- Added explicit preservation cases for ZWJ emoji, bidi prose, and a Unicode identifier.
- Added Unicode 14.0/15.0/15.1 tables; drift errors now list exact missing/extra `U+XXXX` codepoints. This also caught and fixed the supported-Python difference at U+13439–U+1343F.
- Added a reproducible benchmark with warm-up/repeats and separate guard, neutralizer, and complete-preflight scopes.

Verification: focused suite passes on CPython 3.11, 3.12, and 3.13 (`241 passed` each); broader related suite on 3.13: `1010 passed` (one pre-existing thread warning); Ruff and `git diff --check` pass. Latest 480 KiB benchmark: 5.35x guard, 5.16x neutralizer, 5.13x complete request preflight.

efe-arv:
Follow-up re-check complete ✅

I re-ran the maintainer points for PR #98340 against current head `8425d7c06fc89eb3ecc030ff89ab6797eb266c8b` (no scope/code changes needed at this step).

Findings:
- Request preflight still has owner-surface matrix coverage for: instructions, user/assistant content, tool arguments, tool results, and tool schema.
- Unicode table drift guard remains explicit: `_FORMAT_CONTROL_UNICODE_VERSION = unicodedata.unidata_version` + `_FORMAT_CONTROL_PATTERNS` keyed by version + `test_cf_class_matches_unicodedata`.
- Benchmark methodology still reports three separate scopes (guard-only, neutralizer-only, complete preflight) with warmup excluded.

Verification run:
- `uv sync --python 3.11 --extra dev`
- `uv run python -m pytest tests/agent/test_codex_responses_adapter.py -q` -> 241 passed in 4.06s
- `uv run ruff check agent/codex_responses_adapter.py tests/agent/test_codex_responses_adapter.py scripts/benchmark_codex_format_control_scan.py` -> clean
- `uv run python scripts/benchmark_codex_format_control_scan.py --size-kib 64 --warmup 3 --repeats 3 --loops 3` ->
  scope=guard-only old=9.264ms±0.141 new=1.670ms±0.041 speedup=5.55x
  scope=neutralizer-only old=9.549ms±0.065 new=1.753ms±0.021 speedup=5.45x
  scope=complete-request-preflight old=63.014ms±2.461 new=12.371ms±0.333 speedup=5.09x

No new fixes were required, and behavior remains consistent with the previous maintainer-responded change set.

kshitijk4poor:
Thanks for this. It is salvaged into #120406 as part of the perf sweep, with your authorship kept on the commit (cherry-picked with a `cherry picked from` line, or `--author` where it had to be re-applied by hand). The `isascii()` gate is kept. The compiled Cf character class is not used, because it needs a per-Unicode-version table. For non-ASCII text, each distinct character is categorised once instead, which is faster than main on CJK and mixed text.

Closing this in favour of #120406, which is armed to rebase-merge. If that PR doesn't land, I'll reopen this one.

Issue #98338, opened 2026-08-30
auth_native_refresh() splits failures across two logs with zero overlap — 4,927 failures in 17h, 66% invisible in the application log, no backoff or circuit breaker


rlatiolais-afk:
*Version filed against:** `nousresearch/hermes-agent:v2026.8.27` (v0.20.6, upstream 5fc308a7)
**Originally measured on:** v2026.8.19 — every item below re-verified against .27 source on 2026-08-28.
**Environment:** Hosted Fly instance, region sjc, `FLY_VM_MEMORY_MB=4096`. (App name and machine ID redacted from this public issue — available privately to Nous staff on request.)

> Client IPs are redacted: CLIENT-NET-A is the offending client's (NAT'd household) network.

---

## Incident summary

A hosted instance died at 2026-08-26T16:10:43Z under memory pressure. Root-cause analysis over three days established the proximate cause as an operator-initiated resize terminating a VM that was already at capacity — **not** a kernel OOM kill (`oom_kill 0`), not a leak. In the course of that investigation, five defects surfaced, four of them observability failures that materially delayed diagnosis.

A concurrent **17h45m authentication retry storm** (2026-08-26T07:04:55Z → 2026-08-27T00:50:25Z, 3,338 rejected refreshes at up to 3 req/s against `POST https://portal.nousresearch.com/api/oauth/token`) was invisible for two days because of Defect 1. It ended when the client application was removed from the originating machine. Server-side retry behaviour, not the client's bad token, is the defect.

---

### Defect 1 — `auth_native_refresh()` diverges from the dual-logging pattern your own middleware follows at `middleware.py:574/581`
✅ **verified present in v2026.8.27** (`hermes_cli/dashboard_auth/routes.py`, `auth_native_refresh()` at :1028)

Two failure modes, each visible in exactly one log, with zero overlap:

| Exception | agent.log | audit log | HTTP |
|---|---|---|---|
| `RefreshExpiredError` (:1056) | ❌ nothing | ✅ `all_providers_rejected_rt` (:1086) | 401 |
| `ProviderError` (:1058) | ✅ WARNING (:1061) | ❌ nothing | 503 |
| `RefreshExpiredError` via middleware (:569) | ❌ nothing | ✅ `refresh_expired` | — |

[...]

PR #84891, opened 2026-08-12
fix(dashboard-auth): run every blocking IDP provider call off the dashboard event loop

briandevans:
## What does this PR do?

**Symptom:** on a non-loopback bind, the dashboard periodically locks up for
seconds at a time — pages stop loading, the live feed stalls, the terminal
stream freezes — and then everything arrives at once.

**Cause:** the `DashboardAuthProvider` protocol is entirely synchronous
(`hermes_cli/dashboard_auth/base.py:188-209` — every method is a plain `def`),
and the shipped providers implement it with blocking network I/O. All nine
call sites in `hermes_cli/dashboard_auth/` invoke those methods **directly from
`async def` handlers**, so each one blocks the single uvicorn event loop for
its full duration. This PR dispatches all nine through `asyncio.to_thread`.

Three provable ways this bites, in descending order of nastiness:

**(a) A blocking JWKS re-fetch every 300 seconds, bounded at 30.**
`verify_session` → `_verify_jwt` → `PyJWKClient.get_signing_key_from_jwt`. The
client is constructed at `plugins/dashboard_auth/nous/__init__.py:419` and
`plugins/dashboard_auth/self_hosted/__init__.py:600` with
`lifespan=_JWKS_CACHE_SECONDS` (= `300`, at `nous:132` / `self_hosted:127`) and
**no `timeout=`**, so it inherits PyJWKClient's default `timeout = 30`. Its
`fetch_data` is `urllib.request.urlopen(r, timeout=self.timeout, ...)` — a
blocking synchronous socket call. So every 300 seconds the next request to
traverse the auth gate performs a 30s-bounded blocking fetch to the IDP's JWKS
endpoint *on the event loop*, and a slow or blackholed IDP freezes the whole
dashboard for up to half a minute. (In steady state, between those re-fetches,
`verify_session` is a local signature check against a warm cache — this is not
a network round trip on every request.)

**(b) A refresh burst at 10s per call, serialised.** `refresh_session` is never
cached; every call is a token-endpoint round trip at
`_TOKEN_ENDPOINT_TIMEOUT_SEC = 10.0` (`nous:135`, `self_hosted:118`). The
access-token cookie's `Max-Age` tracks the token's own lifetime
(`cookies.py:194` passes the provider-supplied `access_token_expires_in`) while
the refresh-token cookie lives 30 days, so on expiry the browser evicts the AT
and a multi-tab dashboard fires a burst of parallel requests carrying only the
[...]

PR #98342, opened 2026-08-30
fix(dashboard-auth): audit-log provider outages on /auth/native/refresh

chelsealong:
## What does this PR do?

Fixes Defect 1 from #98338: `auth_native_refresh()` (`hermes_cli/dashboard_auth/routes.py:1028`) logs a provider outage (`ProviderError`) only to `agent.log` via `_log.warning(...)`, and never calls `audit_log(...)`. The 503 response it returns leaves zero trace in `dashboard-auth.log`.

This is a real divergence from an established pattern in the same package: `_attempt_refresh` in `hermes_cli/dashboard_auth/middleware.py:573-583` handles the identical `ProviderError` case with **both** a `_log.warning(...)` *and* an `audit_log(AuditEvent.REFRESH_FAILURE, reason="provider_unreachable", ...)` call. `auth_native_refresh` only has the first half of that pattern.

Practical effect (as reported in #98338): during a sustained refresh-token rejection storm against a specific provider, every failure on the native-refresh path (used by the desktop app) is invisible in the audit log — the only signal is a WARNING line in `agent.log`, which is not where operators look for auth failure trends.

## Related Issue

Fixes (partially) #98338 — this addresses Defect 1 only. The issue bundles ~14 separate defects/requests (backoff, circuit breaker, log rotation, memory alerting, etc.); this PR is scoped to the single verifiable logging-divergence bug, matching the "fix real bugs, well" priority in CONTRIBUTING.md rather than attempting the whole report in one patch.

## Type of Change

## Changes Made

- `hermes_cli/dashboard_auth/routes.py`: in `auth_native_refresh()`, add an `audit_log(AuditEvent.REFRESH_FAILURE, provider=..., reason="provider_unreachable", ip=...)` call in the `except ProviderError` branch, mirroring `_attempt_refresh` in `middleware.py`.
- `tests/hermes_cli/test_dashboard_auth_native_flow.py`: add `test_native_refresh_provider_unreachable_is_audited`, which registers a provider whose `refresh_session` raises `ProviderError`, hits `/auth/native/refresh`, and asserts a `provider_unreachable` record lands in `dashboard-auth.log`.

## How to Test

1. `python -m pytest tests/hermes_cli/test_dashboard_auth_native_flow.py -q`
2. Confirm the new test fails without the fix: `git checkout HEAD~1 -- hermes_cli/dashboard_auth/routes.py && python -m pytest tests/hermes_cli/test_dashboard_auth_native_flow.py -q -k provider_unreachable_is_audited` → fails with `AssertionError: provider-unreachable native refresh must be audited` (log file never created). Then `git checkout HEAD -- hermes_cli/dashboard_auth/routes.py` to restore.

### Real output

Before the fix (source reverted to `HEAD~1`, test kept):
```
FAILED tests/hermes_cli/test_dashboard_auth_native_flow.py::test_native_refresh_provider_unreachable_is_audited - AssertionError: provider-unreachable native refresh must be audited
assert False
 +  where False = exists()
 +    where exists = PosixPath('.../.hermes/logs/dashboard-auth.log').exists
1 failed, 1 passed, 14 deselected in 0.89s
```

After the fix:
```
$ python -m pytest tests/hermes_cli/test_dashboard_auth_native_flow.py -q
................                                                         [100%]
16 passed, 7 warnings in 2.33s
```

Full dashboard-auth suite after the fix:
```
$ python -m pytest tests/hermes_cli/test_dashboard_auth_*.py tests/plugins/dashboard_auth tests/plugins/test_plugin_dashboard_auth_contract.py -q
........................................................................ [ 25%]
........................................................................ [ 51%]
........................................................................ [ 77%]
.............................................................            [100%]
277 passed, 7 warnings in 8.74s
```

Lint:
```
$ ruff check hermes_cli/dashboard_auth/routes.py tests/hermes_cli/test_dashboard_auth_native_flow.py
All checks passed!
```

### Code

- [x] I searched for existing PRs to make sure this isn't a duplicate (checked #84891, which touches the same files for an unrelated event-loop-blocking fix, and found no other open PR addressing this audit gap)
- [x] I've run `pytest tests/ -q` (scoped to the dashboard-auth suite, see above) and all tests pass
- [ ] I've tested on my platform — ran in a Linux CI-style container only

- [x] N/A — no config keys, docs, or tool schemas changed

## AI-assistance disclosure

This PR was prepared by an AI coding agent (Claude, via Claude Code) operating autonomously against the public issue tracker. The agent identified Defect 1 in #98338 as the one concretely verifiable, minimal-footprint bug in a much larger multi-part report, verified the divergence against `middleware.py`'s existing dual-logging pattern by reading current `main`, wrote a regression test, and confirmed it fails without the fix and passes with it (see command output above).

### DIFF
--- a/hermes_cli/dashboard_auth/routes.py
+++ b/hermes_cli/dashboard_auth/routes.py
@@ -1062,6 +1062,12 @@ async def auth_native_refresh(request: Request, body: _NativeRefreshBody):
                 "dashboard-auth: provider %r unreachable during native refresh: %s",
                 provider.name, e,
             )
+            audit_log(
+                AuditEvent.REFRESH_FAILURE,
+                provider=provider.name,
+                reason="provider_unreachable",
+                ip=_client_ip(request),
+            )
             continue
         audit_log(
             AuditEvent.REFRESH_SUCCESS,
--- a/tests/hermes_cli/test_dashboard_auth_native_flow.py
+++ b/tests/hermes_cli/test_dashboard_auth_native_flow.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 import hashlib
 import base64
+import json
 import time
 from urllib.parse import parse_qs, urlparse
 
@@ -31,7 +32,7 @@ from hermes_cli.dashboard_auth import (
     register_provider,
 )
 from hermes_cli.dashboard_auth import native_flow
-from hermes_cli.dashboard_auth.base import Session
+from hermes_cli.dashboard_auth.base import ProviderError, Session
 from tests.hermes_cli.conftest_dashboard_auth import StubAuthProvider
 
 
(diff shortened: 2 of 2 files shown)


alt-glitch:

Related to #98344 and #98338: this is the focused provider-unreachable audit fix; #98344 also adds the all-rejected agent-log warning.

liyangbing:
<h3>Review</h3>

<p>This is a focused correction to the native-refresh logging divergence: the provider outage branch now emits <code>REFRESH_FAILURE</code> before continuing, matching the existing middleware path.</p>

<h3>Failure-isolation contract</h3>

<p><a href='https://github.com/NousResearch/hermes-agent/blob/2cff0c1dd835178391367c6bc5cea4a925f1ed67/hermes_cli/dashboard_auth/routes.py#L1062-L1073'>The new audit call is inside the <code>ProviderError</code> branch</a>. Please make audit persistence best-effort for the auth response: if the audit sink is unavailable or raises, the refresh endpoint should still return the intended provider failure status rather than turning an upstream outage into a 500. If <code>audit_log()</code> already guarantees that behavior, add a regression that proves it at this route boundary.</p>

<p>The event currently receives provider name, reason, and client IP, but not the refresh token. Add an assertion that the serialized <a href='https://github.com/NousResearch/hermes-agent/blob/2cff0c1dd835178391367c6bc5cea4a925f1ed67/tests/hermes_cli/test_dashboard_auth_native_flow.py#L589-L642'>audit record</a> contains neither the submitted token nor an authorization header. This is especially important because the test uses a deliberately recognizable token value.</p>

<h3>Regression matrix</h3>

<p>Cover one unreachable provider returning the documented 503 with one audit record, multiple providers where one outage is audited and a later provider can still succeed, audit-writer failure during an outage, repeated refresh failures without token leakage, and a successful refresh proving the existing success event remains unchanged.</p>

<p><strong>Scope:</strong> this review is based on Hermes PR #98342, its stated head revision, and the linked Hermes source/tests. The DeepSeek Harness handbook is an independent community project and is not Hermes documentation or a direct fix, so it is intentionally not linked here.</p>

chelsealong:
Addressed in 7e2c9544f (tests only, no source change needed):

- `audit_log()` already never raises (catches write failures internally, per its docstring), matching the existing pattern in `middleware.py`'s `_attempt_refresh`, which also doesn't wrap the call. Added `test_native_refresh_survives_audit_sink_failure`, which blocks the audit log's directory and confirms `/auth/native/refresh` still returns 503, not 500.
- Extended `test_native_refresh_provider_unreachable_is_audited` to use a recognizable canary refresh token, call the endpoint twice, and assert the token/authorization header never appears in the audit record or raw log across repeated failures.
- Added `test_native_refresh_outage_then_success_both_audited` covering the multi-provider case: one provider's outage is audited, a later provider still succeeds, and the existing `REFRESH_SUCCESS` event is unchanged (provider/user_id present, no token fields).

Full dashboard-auth suite: 279 passed. `ruff check` clean.

PR #55051, opened 2026-06-29
fix(anthropic-oauth): harden Claude Code OAuth runtime restore

Bonchaloo:
## Summary
- Mirrors refreshed Claude Code OAuth credentials into macOS Keychain (existing branch work).
- Prevents restore-time credential-pool reselect from using a stale cross-provider pool, which can poison Anthropic with Codex runtime/base_url state after reload.
- Classifies Anthropic OAuth `out of extra usage` as billing/quota so Hermes can fail over cleanly instead of wedging on a misleading client error.
- Updates curated model catalog with latest Claude picks (`anthropic/claude-fable-5`, `anthropic/claude-opus-4.8-fast`) and regenerated the website catalog.
- Adds Claude Max/OAuth compact prompt mode: skill descriptions become names-only and tool schemas can be deferred behind `tool_search`/`tool_call` (including core tools when configured) so subscription-backed Claude works without Anthropic API credits.

## Verification
- `python3 -m pytest tests/agent/test_restore_primary_pool_reselect.py tests/agent/test_error_classifier.py tests/hermes_cli/test_anthropic_model_flow_stale_oauth.py tests/agent/test_anthropic_keychain.py -q -o 'addopts='` → 203 passed.
- `python3 -m pytest tests/test_tui_gateway_server.py::test_model_options_does_not_overwrite_curated_models -q -o 'addopts='` → passed.
- `npm --workspace apps/desktop run test:ui -- --run src/store/model-visibility.test.ts src/lib/model-status-label.test.ts` → 28 passed.
- `python3 -m pytest tests/tools/test_tool_search.py tests/agent/test_prompt_builder.py -q -o 'addopts='` → 201 passed, 1 skipped.
- Live Claude Code CLI: `claude -p 'Reply OK only.' --model claude-opus-4-8` → OK.
- Live Hermes Anthropic OAuth smoke after compact mode:
  - `hermes chat --cli --provider anthropic -m claude-opus-4-8 -q 'Reply OK only.' -Q --max-turns 1 --source tool` → OK on `claude-opus-4-8`, no fallback.
  - Tool-use smoke with Anthropic OAuth + terminal via `tool_search`/`tool_call` → OK on `claude-sonnet-4-6`, no fallback.

PR #98343, opened 2026-08-30
fix(agent): mirror Claude Code OAuth rotation to the macOS Keychain

### LINKED ISSUE #98334: [Bug]: macOS OAuth refresh writes only ~/.claude/.credentials.json, never the Keychain — stale Keychain refresh token expires the Claude Code login
# [Bug]: On macOS, OAuth refresh writes only `~/.claude/.credentials.json` and never the Keychain — the stale Keychain refresh token bricks the Claude Code login

## Summary

On macOS, Claude Code's authoritative credential store is the **login Keychain** (generic password, service `Claude Code-credentials`). Hermes **reads** that entry but **never writes it**: `git grep add-generic-password -- '*.py'` returns nothing on `main`, while `_write_claude_code_credentials()` writes only the JSON file.

Anthropic's OAuth refresh tokens are **single-use and rotating** — a successful refresh invalidates the previous one. So every Hermes-initiated refresh leaves the Keychain holding an **already-invalidated** refresh token. The next time Claude Code refreshes from the Keychain it gets `invalid_grant`, discards the credential, and the user sees:

```
$ claude auth status --text
Login: Expired — log in again
Not logged in. Run claude auth login to authenticate.
```

Re-logging in fixes it only until the next Hermes refresh, so it recurs indefinitely.

## Not a duplicate of #83338 / #95712

- **#83338** (write-back drops `subscriptionType`) is about *which fields* the file write emits. Fixing field preservation does not help here: the write still never reaches the Keychain, so it stays stale regardless of how complete the file is.
- **#95712** (`suppressed_sources` ignored by the resolver) is about *which identity* is selected. Same relationship — a correct selection still can't repair a Keychain the code never writes.

This issue is the third, independent leg: **on macOS the two stores are written asymmetrically**, so they inevitably diverge. It is also the one that produces the "Claude Code logged itself out" symptom rather than a billing/identity symptom.

The code already knows about the divergence — `read_claude_code_credentials()` (`agent/anthropic_credentials.py:363`) documents it:

```
- If exactly one is non-expired, prefer that one. (Handles the case
  where Claude Code refreshes one source but not the other — observed
  in the wild on Claude Code 2.1.x.)
```

The reconciliation on the read path treats the asymmetry as an external quirk, but the write path is what creates it.

## Root cause

`agent/anthropic_credentials.py`

**Reads both stores** (L363–L399), Keychain included (L281–L284):

```python
kc_creds   = _read_claude_code_credentials_from_keychain()
file_creds = _read_claude_code_credentials_from_file()
```

**Writes one store** — `_write_claude_code_credentials()` (L603):

```python
cred_path = claude_code_credentials_path()   # ~/.claude/.credentials.json
...
os.replace(_tmp_cred, cred_path)
```

Repo-wide, there is no Keychain write:

```console
$ git grep -n "add-generic-password" -- '*.py'
$            # (no output)
```

`_refresh_oauth_token()` POSTs the refresh, rotating the pair, then calls the file-only write. On macOS the Keychain copy is now permanently behind.

## Environment

- Hermes Agent `main` @ `2635035` (2026-08-29) — verified on a clean clone, symptom originally hit on `4209d371a`
- macOS 26.6.2 (Apple Silicon), Python 3.11
- Claude Code CLI 2.1.226, Claude Max subscription
- Hermes `model.provider: anthropic`; no `ANTHROPIC_API_KEY` / `ANTHROPIC_TOKEN` / `CLAUDE_CODE_OAUTH_TOKEN` set, so `resolve_anthropic_token()` reaches the Claude Code branch

## Reproduction

1. macOS, Claude Code CLI logged into a Max/Pro account (both stores populated and in sync).
2. Run Hermes with `provider: anthropic` and no Anthropic credential of its own, so it auto-discovers Claude Code's.
3. Let the access token expire (~8 h, or a reboot) while Hermes is running, then send any Hermes message — Hermes refreshes.
4. Compare the two stores:

```console
$ security find-generic-password -s "Claude Code-credentials" -w \
    | python3 -c 'import json,sys; print(json.load(sys.stdin)["claudeAiOauth"]["refreshToken"][-8:])'
$ python3 -c 'import json,os; print(json.load(open(os.path.expanduser("~/.claude/.credentials.json")))["claudeAiOauth"]["refreshToken"][-8:])'
```

**Expected:** identical (or both refreshed).
**Actual:** they differ — the file has the rotated pair, the Keychain has the invalidated one.

5. Let Claude Code refresh from the Keychain (`claude auth status --text`, or start a session):

**Actual:** `Login: Expired — log in again`. Inspecting the Keychain entry shows Claude Code has emptied it after `invalid_grant`:

```
accessToken:  <EMPTY>
refreshToken: <EMPTY>
expiresAt:    <EMPTY>
scopes:       ['user:file_upload', 'user:inference', 'user:mcp_servers',
               'user:profile', 'user:sessions:claude_code']
subscriptionType: max
```

The metadata survives; only the token triple is cleared. Confirming the token itself was still good at that moment, the file copy answered `HTTP 200` on `POST /v1/messages`.

## Suggested fix

In `_write_claude_code_credentials()`, mirror the write to the Keychain on Darwin when an entry already exists — do not create one where the user has none:

```python
if sys.platform == "darwin" and _read_claude_code_credentials_from_keychain():
    merged = dict(existing_keychain_payload)      # preserve subscriptionType,
    merged.update({                               # rateLimitTier, scopes,
        "accessToken":  access_token,             # refreshTokenExpiresAt
        "refreshToken": refresh_token,
        "expiresAt":    expires_at_ms,
    })
    subprocess.run(
        ["security", "add-generic-password", "-U",
         "-a", getpass.getuser(), "-s", "Claude Code-credentials",
         "-w", json.dumps({"claudeAiOauth": merged})],
        check=True,
    )
```

Notes for whoever picks this up:

- `-U` updates the existing item in place; without it `security` errors on a duplicate.
- Merge into the existing payload rather than replacing it. Claude Code ≥ 2.1.81 gates on `user:inference` being present in the stored `scopes`, and `subscriptionType` drives Max/Pro vs API-key mode (the #83338 failure mode — worth fixing consistently on both stores).
- Same TOCTOU care as the file path: the payload is a secret and `-w <value>` puts it on argv, where it is visible in the process table. `security` prompts for the password interactively when `-w` is given no value, so a non-argv path may be available — worth checking before merging.
- A regression test would assert that after a refresh on Darwin, both stores return the *same* `refreshToken` — an invariant, not a snapshot.

Happy to send a PR if the approach looks right. One design question worth settling first: whether writing another app's Keychain item is acceptable at all, or whether Hermes should instead **stop refreshing auto-discovered Claude Code credentials** and require its own (`hermes auth add anthropic`) — which is closer to the direction in #82235 and sidesteps the shared single-use-token lineage entirely.

## Workaround

Two steps are needed; either alone is insufficient.

1. Give Hermes its own credential so it stops depending on Claude Code's:
   ```console
   $ hermes auth add anthropic --type oauth --label hermes-own
   $ hermes auth remove anthropic claude_code
   ```
2. `hermes auth remove` is **not** enough on its own — that is #95712: `suppressed_sources` gates only pool seeding, while the resolver still reads the file directly. Until that lands, the resolver order has to be flipped locally so the pool credential wins (`agent/anthropic_credentials.py:834` vs `:839` — move the credential-pool branch above the Claude Code branch).

With both applied, five consecutive `resolve_anthropic_token()` calls left `~/.claude/.credentials.json` byte-identical (mtime unchanged) and `claude -p` returned `"subtype":"success"`.

To recover a Keychain that has already been emptied without a fresh browser login: if the file copy is still live, merge its token triple back into the existing Keychain payload (preserving `scopes` / `subscriptionType`) with `security add-generic-password -U`.

liuhao1024:
## What does this PR do?

On macOS, Claude Code's authoritative credential store is the login Keychain (generic password, service `Claude Code-credentials`). Hermes reads that entry but never wrote it: `_write_claude_code_credentials()` only updated `~/.claude/.credentials.json`. Since Anthropic OAuth refresh tokens are single-use and rotating, every Hermes-initiated refresh stranded a **spent** refresh token in the Keychain; Claude Code's next refresh then failed with `invalid_grant` and emptied the entry (`Login: Expired — log in again`), recurring until the user re-logged in.

This PR makes `_write_claude_code_credentials()` mirror the committed rotation to the Keychain on Darwin:

- **Existing entries only** — `_read_claude_code_keychain_entry()` reads the item's real `acct` attribute (rather than assuming the login user name) and the raw `claudeAiOauth` payload; if no entry exists, the store is left untouched, so a user whose Claude Code install doesn't use the Keychain never gains one. An unparsable `acct` also stands down: `add-generic-password -U` matches on `(service, account)`, and a wrong guess would create a duplicate entry next to the one Claude Code reads.
- **Merge, not replace** — only the rotated token fields are merged into the stored payload, so the metadata Claude Code gates on (`scopes` incl. `user:inference`, `subscriptionType`, `rateLimitTier`, ...) survives the rotation, including the post-`invalid_grant` state where Claude Code has already emptied the token triple.
- **Secret stays off argv** — the update goes through `security -i` (command parsed from stdin) instead of `-w <value>` on the command line. The escaping (`\` and `"` only; single quotes are ordinary characters to this parser — shell-style `'..'` spans silently truncate at an embedded quote) is verified against the real CLI and covered by `_security_i_escape()`'s unit test plus a mocked round-trip assertion.
- **Best-effort after the commit** — the JSON file write remains the commit step of the refresh transaction; every caller (the direct resolver path and both `credential_pool` paths) treats an exception from this function as a failed rotation and would mark an already-committed rotated pair as spent. A locked/unavailable Keychain therefore logs an error (naming the recovery: `claude login`) instead of un-committing the rotation.

## Related Issue

Fixes #98334

## Type of Change

## Changes Made

- `agent/anthropic_credentials.py`
  - Extracted the shared `_CLAUDE_CODE_KEYCHAIN_SERVICE` constant (previously the service name was inline in the reader).
  - Added `_read_claude_code_keychain_entry()`: reads `(account, raw claudeAiOauth payload)` from the existing entry, with `acct` parsed from item metadata so `add-generic-password -U` matches the entry Claude Code actually owns.
  - Added `_security_i_escape()` and `_sync_claude_code_keychain_credentials()`: the mirror itself — Darwin-gated, existing-entry-only, token-fields-only merge, `security -i` stdin transport, best-effort error handling with an actionable log message.
  - `_write_claude_code_credentials()` calls the sync after the atomic file replace succeeds.
- `tests/conftest.py` — extended the `_neutralize_macos_keychain_creds` autouse fixture to also neutralize `_read_claude_code_keychain_entry`, so ordinary credential-file tests on a macOS host cannot touch (or rotate) the developer's real Keychain entry.
- `tests/agent/test_anthropic_keychain.py` — regression tests (details under How to Test).

## How to Test

1. `python -m pytest tests/agent/test_anthropic_keychain.py -q` → **17 passed, 1 skipped** (the `linux_only` class skips on macOS; it asserts zero `security` calls off Darwin and runs for real on the Linux lane).
2. `python -m pytest tests/agent/test_anthropic_adapter.py tests/agent/test_anthropic_credential_persist_failure.py tests/agent/test_anthropic_oauth_stress.py tests/agent/test_anthropic_borrowed_row_authority.py tests/agent/test_credential_pool_anthropic_refresh_race.py tests/agent/test_auxiliary_client.py -q` → **300 passed, 1 skipped**.
3. `python -m pytest tests/agent/ -q -k "credential or anthropic or keychain or oauth"` → **658 passed, 6 skipped, 1 pre-existing failure** (`test_review_engine.py::test_delegate_task_credentials_cfg_overrides_delegation_config` fails identically on clean `upstream/main`, unrelated to this change).
4. End-to-end against the real `security` CLI (throwaway service name, never the real entry): after `_sync_claude_code_keychain_credentials()` on an entry whose payload contains quotes/backslashes, `security find-generic-password -s <svc> -w` round-trips byte-identical JSON with the token triple rotated and `subscriptionType` preserved. Observed result: `E2E roundtrip OK`.

Key regression assertions (mocked `security`, no real Keychain touched):
- after a write, both stores hold the same rotated `refreshToken` and the merge preserves `subscriptionType`/`rateLimitTier`/`scopes`;
- the payload never appears in any subprocess argv (stdin transport);
- no Keychain entry → no `security add-generic-password` call at all (file still commits);
- emptied post-`invalid_grant` entry → mirror still repopulates the token triple and keeps metadata;
- `security -i` non-zero exit or `TimeoutExpired` → the committed file write still succeeds (no `CredentialPersistError`).

### Code

- [x] I've run `pytest tests/ -q` and all tests pass *(targeted credential/anthropic/keychain/oauth selection: 658 passed; the single failure is pre-existing on clean main, shown above)*
- [x] I've tested on my platform: macOS 26.6.2 (Apple Silicon, Darwin arm64), Python 3.14

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — docstrings document the mirror contract and the `security -i` escaping rationale
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — Darwin-gated; Linux lane asserts zero `security` calls; Windows unaffected (the reader already returned `None` off Darwin)

### DIFF
--- a/agent/anthropic_credentials.py
+++ b/agent/anthropic_credentials.py
@@ -28,6 +28,7 @@ import json
 import logging
 import os
 import platform
+import re
 import secrets
 import stat
 import subprocess
@@ -262,6 +263,9 @@ def is_rotation_consumed_uncommitted(
     return fingerprint in _read_spent_rotation_sidecar(source_path)
 
 
+_CLAUDE_CODE_KEYCHAIN_SERVICE = "Claude Code-credentials"
+
+
 def _read_claude_code_credentials_from_keychain() -> Optional[Dict[str, Any]]:
     """Read Claude Code OAuth credentials from the macOS Keychain.
 
@@ -281,7 +285,7 @@ def _read_claude_code_credentials_from_keychain() -> Optional[Dict[str, Any]]:
         # Read the "Claude Code-credentials" generic password entry
         result = subprocess.run(
             ["security", "find-generic-password",
-             "-s", "Claude Code-credentials",
+             "-s", _CLAUDE_CODE_KEYCHAIN_SERVICE,
              "-w"],
             capture_output=True,
             text=True, encoding='utf-8', errors='replace',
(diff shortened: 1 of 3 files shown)


alt-glitch:

Duplicate of #55051: both add the missing macOS Keychain write-back after Claude Code OAuth token rotation.

liuhao1024:
The triage flag is correct — closing in favor of #55051, which already implements the same macOS Keychain write-back after Claude Code OAuth token rotation (`_write_claude_code_credentials_to_keychain` via `security add-generic-password -U`) and predates this PR by two months.

Kept the branch (`liuhao/cron-bugfix-98334`) on my fork in case a maintainer would rather land the keychain fix as a focused change instead of as one concern inside the larger PR — happy to reopen if that's preferred. Thanks @alt-glitch for the catch.

Issue #89896, opened 2026-08-19
[Bug]: Desktop auto-speak still races after 63565fa26 — now doubles, truncates, or skips


darrinmcgann:
### Summary

`63565fa26` — *fix(desktop): keep auto-speak silent across the stream-id rewrite* (supersedes #75649, #86637, #87672, #88642; closes #86601, #87652, #87823) — does not resolve the auto-speak race on Windows. It changes **how** it fails rather than eliminating it: in addition to the original double-playback, two new failure modes appear (total silence, and truncation mid-reply).

All three modes were observed **in a single session, on a single process, running the fixed bundle**, on consecutive replies.

### Environment

| | |
|---|---|
| OS | Windows 10 |
| Surface | Desktop app (Electron) |
| Version | v0.20.4 (2026.8.18) |
| HEAD | `63565fa26b00a2096247064785c4380aafab2303` |
| Install method | git |
| Python | 3.11.16 |
| TTS provider | OpenAI (`gpt-4o-mini-tts`, voice `alloy`) via Nous Portal |
| Model | `anthropic/claude-opus-5` (provider `nous`) |
| `voice.auto_tts` | `true` |
| Input-bar speaker icon | enabled |

### Verified running the fixed code

Electron maps `app.asar` at process start, so this was checked explicitly to rule out a stale bundle:

```
fix commit 63565fa26   2026-08-19 03:56:55
app.asar built         2026-08-19 04:33:16   (newer than commit)
Hermes.exe started     2026-08-19 04:34:59   (newer than bundle)
```

Ordering is correct — the running renderer contains the fix.

### Observed failure modes

Consecutive assistant replies, same process, no config changes between them:

1. **Doubled** — full reply spoken twice from the top
2. **Silent** — nothing played (backend generated audio normally)
3. **Truncated** — roughly half the reply spoken, then cut off mid-stream

The alternation between modes is what identifies this as a race rather than a deterministic fault.

### Backend is not implicated

`tools.tts_tool` logs an identical, complete generate→save cycle for all three outcomes:

```
04:38:09,822 INFO tools.tts_tool: Generating speech with OpenAI TTS...
04:40:03,446 INFO tools.tts_tool: Generating speech with OpenAI TTS...
04:41:04,229 INFO tools.tts_tool: Generating speech with OpenAI TTS...
04:42:43,657 INFO tools.tts_tool: Generating speech with OpenAI TTS...
```

[...]

PR #98346, opened 2026-08-30
fix(desktop): keep durable spoken replies consumed

0jarvis1337-afk:
## What does this PR do?

Prevents Hermes Desktop auto-speak from replaying a durable assistant reply after a stale mounted message snapshot moves the session's spoken anchor backward.

The existing live-to-durable rewrite logic handles a streaming reply changing IDs. A separate race remains when an older mounted snapshot later calls `markAssistantIdSpoken`: the single stored anchor can regress to an earlier durable reply. When playback returns to idle, the current snapshot can then present the already-spoken reply as pending and queue it again.

This change keeps a small, session-scoped, insertion-ordered history of durable reply IDs that have already been consumed. `resolveSpokenReply` uses that history to avoid regressing behind the newest consumed durable reply present in the caller's snapshot. Live-tail IDs remain governed by the existing rewrite/ordinal behavior, so compaction can still establish a new low-ordinal live anchor.

Related: #89896. This targets the repeat-from-the-top mode reported there; it does not claim to address the issue's separate silent or truncated playback modes.

This overlaps symptomatically with #91391, but addresses a different state transition: durable-anchor regression from stale mounted snapshots rather than filtering live/interim candidates.

## Type of Change

- [x] Bug fix (non-breaking change that fixes incorrect behavior)

## Changes Made

- Add a bounded history of recently consumed durable assistant IDs per session.
- Prefer the newest consumed durable reply visible in a stale snapshot without moving the live frontier backward.
- Preserve existing live-to-durable rewrite behavior.
- Add regression coverage for stale snapshots, repeated notifications, transcript compaction, later distinct replies, and live-to-durable rewrites.

## How to Test

From `apps/desktop`:

```bash
npx vitest run --project ui src/lib/spoken-reply.test.ts src/app/chat/composer/hooks/use-auto-speak-replies.test.tsx
npm run typecheck
npx eslint src/lib/spoken-reply.ts src/lib/spoken-reply.test.ts
npm run build
```

All commands pass on macOS (Apple Silicon). The targeted Vitest run passes 14 tests.

## Checklist

- [x] I read the contributing guide.
- [x] I searched open issues and pull requests for duplicates/adjacent fixes.
- [x] The commit uses Conventional Commits.
- [x] The PR is narrowly scoped to the Desktop spoken-reply selector.
- [x] Regression tests are included.
- [x] Desktop typecheck, focused lint, targeted tests, and production build pass.
- [x] Cross-platform impact considered: renderer-only TypeScript; no native or platform-specific APIs.
- [x] No configuration keys or documentation behavior changed.

### DIFF
--- a/apps/desktop/src/lib/spoken-reply.test.ts
+++ b/apps/desktop/src/lib/spoken-reply.test.ts
@@ -89,4 +89,55 @@ describe('resolveSpokenReply', () => {
     expect(assistantReplyOrdinal(nextTurn, 'assistant-stream-2')).toBe(1)
     expect(spoken?.ordinal).toBe(0)
   })
+
+  it('uses durable history when a stale snapshot marks an older durable reply', () => {
+    const current = [assistant('durable-1'), assistant('durable-2')]
+    markAssistantIdSpoken('s', current, 'durable-2')
+
+    const stale = [assistant('durable-1')]
+    markAssistantIdSpoken('s', stale, 'durable-1')
+
+    expect(resolveSpokenReply('s', stale)).toEqual({ id: 'durable-1', ordinal: 0 })
+    expect(resolveSpokenReply('s', current)).toEqual({ id: 'durable-2', ordinal: 1 })
+  })
+
+  it('lets a new low-ordinal live reply become consumed after compaction', () => {
+    const beforeCompaction = [
+      assistant('durable-1'),
+      assistant('durable-2'),
+      assistant('durable-3')
+    ]
+
+    markAssistantIdSpoken('s', beforeCompaction, 'durable-3')
+
+    const compacted = [assistant('assistant-stream-new')]
+    markAssistantIdSpoken('s', compacted, 'assistant-stream-new')
+
+    expect(spokenReplyOf('s')).toEqual({ id: 'assistant-stream-new', ordinal: 0 })
+    expect(resolveSpokenReply('s', compacted)).toEqual({ id: 'assistant-stream-new', ordinal: 0 })
+  })
+
+  it('keeps repeated notifications for the same durable reply consumed', () => {
+    const messages = [assistant('durable-1')]
+    markAssistantIdSpoken('s', messages, 'durable-1')
+    markAssistantIdSpoken('s', messages, 'durable-1')
+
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related to #91391 and #89896. This PR addresses stale durable-anchor regression, rather than the live/interim filtering and stream-ID hydration paths covered by those fixes.

liyangbing:
<h3>Review</h3>
<p>A bounded session-scoped consumed-ID history is a sensible fence against stale mounted snapshots replaying a durable spoken reply. The key invariant is monotonic consumption without confusing durable identity with the live-tail frontier.</p>
<h3>Suggested contract</h3>
<ul>
<li>Record consumption against a stable durable reply identity plus session generation; a stale snapshot must never consume a reply from a newer session or profile generation.</li>
<li>Make repeated notifications idempotent and keep the bounded history eviction policy explicit. Eviction must not make an already-consumed durable reply eligible merely because a later snapshot is old.</li>
<li>Keep the live-to-durable rewrite mapping separate from the consumed set, and invalidate both on an explicit session reset or compaction generation change.</li>
<li>Persist or reconstruct the consumed frontier consistently with the durable transcript so a renderer restart cannot replay audio that the previous process already acknowledged.</li>
<li>Expose redacted reason codes for skipped, stale, rewritten, and newly queued replies; do not infer playback success from selection alone.</li>
</ul>
<h3>Regression matrix</h3>
<ul>
<li>stale mounted snapshot, repeated notification, durable ID rewrite, compaction, and a later distinct reply</li>
<li>renderer restart before and after playback acknowledgment, session switch, and profile or generation change</li>
<li>bounded-history eviction, duplicate IDs, missing IDs, out-of-order ordinals, and a live-tail reply</li>
<li>playback failure, cancellation, retry, and concurrent idle transitions, verifying selection and actual consumption remain separate</li>
</ul>
<p>The handbook's <a href="https://sandbaseai.github.io/deepseek-harness-handbook/session-history-recovery.html">Session-history recovery guide</a> and <a href="https://sandbaseai.github.io/deepseek-harness-handbook/session-log-durability.html">Session-log durability guide</a> cover the same generation, replay, and acknowledgement boundaries.</p>

PR #98347, opened 2026-08-30
fix(guard): support multipart list content in continuation repetition guard

### LINKED ISSUE #86581: [Bug]: Truncated-response continuation (finish_reason=length) never checks for repetition before retrying, and Discord delivery has no split-count cap — one degenerate turn floods a channel with 30+ messages
### Bug Description

A single agent turn produced a final response of 60,698 characters, delivered to a live Discord channel as **31 separate messages** posted back-to-back. Reconstructed from `state.db` message rows and gateway logs, this happened in two independently-missing safety checks, not one:

1. **Generation side**: the model entered degenerate repetition mid-response, hit the provider's output-token ceiling (`finish_reason=length`), and Hermes's own truncation-continuation path retried it — appending a "continue, don't repeat" nudge — **without ever checking whether the content it was about to continue was already pathological**.
2. **Delivery side**: the resulting 60,757-char combined response was hard to miss (Discord's own 2000-char limit forces a split), yet the platform adapter splits and sends **every** resulting chunk with no upper bound on count.

Either check alone would have prevented the user-facing flood.

### Root Cause (confirmed against `main` @ `76d832d3`, Hermes Agent v0.20.0 / 2026.8.3, cross-checked against `state.db` message rows for the affected session)

**1. (Primary) `finish_reason=length` triggers an automatic continuation with zero content-sanity check on the truncated text.**

Evidence from `state.db` (`messages` table, `session_id='20260814_142129_1a77534b'`):

| id | role | len | finish_reason | timestamp (UTC) |
|---|---|---|---|---|
| 39020 | assistant | 57,752 | `length` | 14:56:05.907 |
| 39021 | user (synthetic) | 176 | — | 14:56:05.907 (same instant) |
| 39036 | assistant | 2,945 | `stop` | 14:58:24.877 |

`57752 + 2945 = 60697` ≈ the `60698`-char final response the gateway delivered — the two fragments were stitched into one turn.

Message 39020's content is not random garbage — it opens with `"We need to verify the model setting. The user wants ... Then '...' Then '...'"`: the model was narrating/summarizing the conversation so far, one prior user turn at a time. When it reached the point in history where the user had once said *"好，你幫我更改成 Google Gemini 4 31B。"* (said exactly **once**, at 14:33:33 — confirmed by scanning all 16 user-role rows in this session; there is no duplicate inbound message), it entered a token-level repetition loop echoing that fragment, verbatim, hundreds of times, consuming the rest of its output budget until the provider cut it off.

Message 39021 is Hermes's own auto-injected continuation nudge — `agent/conversation_loop.py:781-785` (`_LENGTH_CONTINUATION_OUTPUT_LIMIT`):
```python
_LENGTH_CONTINUATION_OUTPUT_LIMIT = (
    "[System: Your previous response was truncated by the output "
    "length limit. Continue exactly where you left off. Do not "
    "restart or repeat prior text. Finish the answer directly.]"
)
```
This fires from the `finish_reason == "length"` branch at `agent/conversation_loop.py:3196`. The branch already contains a specific, targeted sanity check for a *different* pathological case — thinking-budget exhaustion (`_thinking_exhausted`, `conversation_loop.py:3196-3260`) — which aborts immediately with a user-facing `"⚠️ Thinking Budget Exhausted"` message instead of retrying. **No equivalent check exists for "the truncated content is dominated by repeated text."** The continuation path (`conversation_loop.py:3369-3412`, guarded only by `length_continue_retries < 4` at line 1576/3333) unconditionally appends the truncated fragment to `messages` and asks the model to keep going. In this incident it only needed 1 of the 4 available retries — the second pass finished cleanly (`finish_reason=stop`, no further repetition) — so the retry-count ceiling never engaged; nothing else was in a position to catch the first fragment's content quality before it was folded into the final response.

**2. Delivery side: the chunk-send loop has no count/size ceiling either.**
`plugins/platforms/discord/adapter.py:1002-1003`:
```python
MAX_MESSAGE_LENGTH = 2000
_SPLIT_THRESHOLD = 1900  # near the 2000-char split point
```
`plugins/platforms/discord/adapter.py:3114-3122`:
```python
# Format and split message if needed
formatted = self.format_message(content)
chunks = self.truncate_message(formatted, self.MAX_MESSAGE_LENGTH)

message_ids = []
reference = self._reply_reference_for_send(reply_to, channel)

for i, chunk in enumerate(chunks):
```
`chunks` is iterated in full regardless of `len(chunks)`. No `if len(chunks) > N` guard exists here, in the structurally identical `_send_to_forum` path (line 3206), or the edit-overflow path (line 3473). A 2-chunk reply and a 200-chunk reply take the exact same code path.

**3. The gateway's own split-delivery bookkeeping already *knows* a split is happening, but only uses that for a duplicate-suppression decision, not a circuit breaker.**
`gateway/run.py:27059` (this incident's actual log line):
```
Stale streamed finalize detected for session %s on a multi-message split; skipping the in-place reconciliation edit and delivering the complete response via normal final send (#78541).
```
`_turn_split_delivery` (`gateway/run.py:27050-27059`) is a concrete, checked signal that the current response is being delivered as a multi-message split, available *before* the send happens. It is used only to pick edit-vs-fresh-send; never to ask "is this split unreasonably large?"

**4. Response-shape validation exists but only checks structural validity, not content sanity.**
`agent/conversation_loop.py:2786-2864` — `validate_response()` (dispatched per `api_mode`) checks for `response is None`, missing/empty `choices`, missing `content`, etc. A structurally well-formed response with `finish_reason=stop` and 60,698 chars of repeated text passes every check here.

**5. Contributing factor — the turn ran to 108K tokens across 10 API calls with no early exit.** Before producing the oversized answer, the agent spent ~15 minutes retrying blocked tool calls (`hermes model` requiring a TTY; an internal guard rejecting self-restart from inside the gateway process). Nothing in the retry/iteration accounting (turn ended at `api_calls=10/90`, well under budget) treats "many consecutive tool failures without task progress" as a signal to stop and hand back a short, controlled error instead of continuing to grow context — the polluted, 100K+-token context is the substrate the repetition loop in #1 happened on top of.

**6. Related, lower-confidence observation — not the cause of this incident, but same risk class.** `agent/background_review.py:3` documents that after most turns, `AIAgent.run_conversation` may fork a background self-review agent that replays the **full, unfiltered parent conversation snapshot** on the same model/runtime. `agent/background_review.py:757-758` shows the fork only inherits a `max_tokens` cap if the *parent* explicitly set one — no independent ceiling otherwise. In this incident the review fork for the same session was triggered 1.4s after the flood turn ended and was itself killed mid-call by a gateway restart (`Broken pipe` in the logs) before it produced anything — so it did not cause this flood, but it inherited the exact same 108K-token polluted snapshot and had no independent output-length guard either. Flagging as a related gap, not the root cause of this ticket.

### Steps to Reproduce

1. Get a model into a state where, while narrating/summarizing prior conversation turns, it echoes one fragment repeatedly (long, error-laden context measurably increases the odds — see #5 above — but this is a model-degeneracy trigger, not a Hermes-specific one).
2. Let generation run to the provider's output-token ceiling (`finish_reason=length`) while still inside the repeated fragment.
3. Observe: `conversation_loop.py`'s length-continuation path retries automatically with no inspection of the truncated content (root cause #1).
4. Whether or not the continuation itself repeats, the stitched final response can already be tens of thousands of characters. Observe: the Discord adapter splits and sends **all** resulting chunks with no ceiling (root cause #2).

Root cause #2 alone is trivially reproducible without reproducing model degeneracy: call `DiscordAdapter.send()` with any string >60,000 chars against a live channel and count outbound messages.

### Actual Behavior

```
state.db: id=39020 role=assistant len=57752 finish_reason=length  (14:56:05.907 UTC)
state.db: id=39021 role=user     len=176   [System: ...truncated... Continue...] (same instant, synthetic)
state.db: id=39036 role=assistant len=2945  finish_reason=stop     (14:58:24.877 UTC)

gateway.run: response ready: platform=discord chat=1537828505927033063 time=882.3s api_calls=10 response=60698 chars
[Discord] Sending response (60757 chars) to 1537828505927033063
→ 31 individual Discord messages posted back-to-back
```

### Expected Behavior

1. In the `finish_reason == "length"` branch (`conversation_loop.py:3196`), before appending the continuation nudge, run a cheap content-sanity check on the truncated fragment (e.g. n-gram repetition ratio, or simply: longest exact substring repeated >K times) analogous to the existing `_thinking_exhausted` check at the same call site. On a pathological result, abort the turn with a short, user-facing error (same pattern as the existing `"⚠️ Thinking Budget Exhausted"` message) instead of continuing.
2. Independently, `DiscordAdapter.send()` (and the identical `_send_to_forum` / edit-overflow paths) should cap the number of chunks sent for one logical response — e.g. a configurable `max_split_messages` (suggested default: 5-8) — falling back to a short "response was abnormally large, see logs" notice instead of flooding.
3. The existing `_turn_split_delivery` bookkeeping in `gateway/run.py` already has the chunk context needed to enforce (2) platform-agnostically at the gateway layer instead of duplicating it per-adapter.

### Proposed Fix (sketch, for a follow-up PR)

1. Add a bounded repetition check (e.g. `agent/repetition_guard.py`, called from `conversation_loop.py:3196` right after `_thinking_exhausted` is ruled out) gating the length-continuation path.
2. Add `DiscordAdapter.MAX_SPLIT_MESSAGES` (or a shared constant in `gateway/platforms/base.py`) checked right after `chunks = self.truncate_message(...)` at `adapter.py:3116`, `3206`, `3473`.
3. Optionally surface `_turn_split_delivery`'s chunk count in `gateway/run.py` before calling adapter `send()` so (2) can be enforced once, platform-agnostically.

### Environment

- **Hermes Agent**: v0.20.0 (2026.8.3) · upstream commit `76d832d3857551a029c4b39c23945eb47c16fe5b`
- **Install method**: Docker (`nousresearch/hermes-agent:latest`)
- **Platform**: Discord (guild channel, non-forum, non-thread — plain `channel.send()` path)
- **Provider/model active during incident**: `nvidia` / `nvidia/nemotron-3-nano-omni-30b-a3b-reasoning` (verified against a same-day NVIDIA `/v1/models` snapshot — model ID itself was valid; this is not a "model doesn't exist" report)
- Verified via direct log + `state.db` inspection against the exact running commit (matched `hermes version` output to `git log -1` in a full clone of upstream at that commit).

### Related Work

- **#30408** — "Hermes should stop repeated tool loops and block obviously broken final responses." Closed 2026-06-28 as `completed`, **no linked PR**, maintainer (`teknium1`) declined the model-behavior framing verbatim: *"We don't want to fix models problems. There are also plenty of legitimate paths to repeat tool calls."* This report deliberately does not re-ask for model-output judgment. Root causes #1-#4 below are Hermes's own retry/validation/delivery code, independent of whether the model "should" have produced that output — the same distinction the maintainer's own closed **#10711** (below) already accepted for a structurally identical case.
- **#10711** — "Ollama returns `finish_reason='stop'` on truncated GLM responses, causing agent to silently drop final output" (closed, `completed`). Established the precedent that Hermes must not blindly trust a provider's finish-signal and needs its own content-side verification before treating a response as final. Root cause #1 in this report asks for the same class of check on the *other* side of that precedent: `finish_reason=length` is correctly detected, but the continuation path still doesn't inspect the truncated content before retrying.
- **#13208** (open, unresolved) + **#6784** / **#8528** (open PRs, unmerged: "tool loop detection with context pruning" / "integrate tool loop detection into agent loop") — cover contributing factor #5 (no-progress tool-call loops growing context unchecked). This report does not duplicate that ask; it references it only as background for why the turn reached 108K tokens before the repetition in #1 occurred. If #6784/#8528 land, they may reduce (but would not eliminate) exposure to #1 and #2 — the content-check and delivery-cap gaps are independent of whether the loop that grew the context gets shortened.
- **#78541** — "Suppressing normal final send swallows complete replies on Telegram group/forum sessions" (closed, fixed via #79669) — established the `_turn_split_delivery` flag this incident's payload traveled through. That fix addressed a *suppression* bug (legitimate content silently dropped); this report is the opposite direction (oversized content delivered with no ceiling) through the same code path.
- **#71643** — "Telegram streaming — successful finalize edit carries the stale preview text" (closed) — the stale-finalize reconciliation-edit fallback referenced in the same `gateway/run.py` block.
- **#25349** — "Discord streaming can duplicate final responses across chunked delivery" (open) — same general code area, different symptom (duplication vs. unbounded volume).
- **#75423** (open) — "`_SKILL_REVIEW_PROMPT` turns one-off investigations into documentation-shaped Skills" — a different symptom of the same background-review mechanism flagged as a related risk in root cause #6; corroborates that the review fork has more than one open edge case around what context/output it's trusted with.

### Additional Context

- No pending/queued outbound messages were involved — this was a single turn's response, delivered once, not a retry loop re-sending old content.
- The incident coincided with, but was not caused by, a global model configuration change and gateway restart the user performed shortly after (switching to `nvidia/nemotron-3.5-lightning-30b-a3b`, unrelated to the model active during the flood). Nothing in the user-facing output distinguished "this flood came from a turn already in flight" from "this flood came from the model you just switched to" — worth a documentation/UX note even outside this specific bug.

ericmaddox:
## What does this PR do?

Fixes a **P0 (Severe Denial of Service / Message Flooding / Infinite Loop Continuation)** bug in `agent/repetition_guard.py` and `agent/conversation_loop.py` where truncated responses (`finish_reason=length`) containing multipart or multimodal content blocks (`[{"type": "text", "text": "..."}]`) completely bypassed the repetition guard because `is_repetition_dominated()` only accepted `str` and returned `False` on list inputs.

In Hermes Agent:
1. When a model output enters a pathological repetition loop and truncates at `finish_reason=length`, the repetition guard detects the degenerate repetition and aborts the turn cleanly instead of appending a continuation nudge that would stitch dozens of repetitive turns into channels (issue #86581).
2. For models returning structured multipart lists (standard across Anthropic, Gemini, Bedrock, and OpenAI Responses API), `_trunc_content` is a `list`.
3. `re.search` for `<think>` tags crashed with `TypeError: expected string or bytes-like object, got 'list'`, and `is_repetition_dominated()` returned `False` immediately, completely breaking repetition protection for structured responses.
4. The fix ensures:
   - `_trunc_text` is cleanly derived using `flatten_message_text()` when `_trunc_content` is a list, enabling safe think-tag search and visible text extraction.
   - `is_repetition_dominated()` flattens multipart lists before evaluating repetition dominance.
   - Unit tests added in `tests/agent/test_repetition_guard.py` and `tests/run_agent/test_continuation_repetition_guard.py`.

## Related Issue

Fixes #86581 (multipart/multimodal edge case)

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `agent/repetition_guard.py`:
  - Accepted multipart list content in `is_repetition_dominated()` by flattening with `flatten_message_text()`.
- `agent/conversation_loop.py`:
  - Safely extracted `_trunc_text` from string or list content for thinking-tag search and repetition detection.
- `tests/agent/test_repetition_guard.py`:
  - Added unit test verifying multipart lists with repeated text are flagged.
- `tests/run_agent/test_continuation_repetition_guard.py`:
  - Added E2E conversation loop test verifying continuation aborts when multipart list response is repetition-dominated.

## How to Test

Run the repetition guard test suites:
```bash
uv run --with pytest python -m pytest tests/agent/test_repetition_guard.py tests/run_agent/test_continuation_repetition_guard.py -v
```

### Code

- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(guard): support multipart list content in continuation repetition guard`)
- [x] I've tested on my platform: Windows 11 / Linux

## Screenshots / Logs

```text
============================= test session starts =============================
platform win32 -- Python 3.11.15, pytest-9.1.1, pluggy-1.6.0
rootdir: hermes-agent
configfile: pyproject.toml
plugins: anyio-4.12.1
collected 11 items

tests/agent/test_repetition_guard.py ........                            [ 72%]
tests/run_agent/test_continuation_repetition_guard.py ...                [100%]

============================= 11 passed in 13.81s =============================
```

### DIFF
--- a/agent/conversation_loop.py
+++ b/agent/conversation_loop.py
@@ -95,6 +95,7 @@ from agent.retry_utils import (
     jittered_backoff,
     zai_coding_overload_retry_ceiling,
 )
+from agent.message_content import flatten_message_text
 from agent.repetition_guard import is_repetition_dominated
 from agent.trajectory import has_incomplete_scratchpad
 # Bind before the turn starts so a source-tree swap cannot load a skewed
@@ -3757,6 +3758,11 @@ def run_conversation(
                     _trunc_msg = _trunc_result
 
                     _trunc_content = getattr(_trunc_msg, "content", None) if _trunc_msg else None
+                    _trunc_text = (
+                        flatten_message_text(_trunc_content)
+                        if isinstance(_trunc_content, list)
+                        else (_trunc_content if isinstance(_trunc_content, str) else "")
+                    )
                     _trunc_has_tool_calls = bool(getattr(_trunc_msg, "tool_calls", None)) if _trunc_msg else False
 
                     # ── Detect thinking-budget exhaustion ──────────────
@@ -3772,9 +3778,9 @@ def run_conversation(
                     # truncations that deserve continuation retries, not as
                     # thinking-budget exhaustion.
                     _has_think_tags = bool(
-                        _trunc_content and re.search(
+                        _trunc_text and re.search(
                             r'<(?:think|thinking|reasoning|REASONING_SCRATCHPAD)[^>]*>',
-                            _trunc_content,
+                            _trunc_text,
                             re.IGNORECASE,
                         )
                     )
@@ -3782,8 +3788,8 @@ def run_conversation(
                         not _trunc_has_tool_calls
                         and _has_think_tags
                         and (
-                            (_trunc_content is not None and not agent._has_content_after_think_block(_trunc_content))
-                            or _trunc_content is None
+                            bool(_trunc_text and not agent._has_content_after_think_block(_trunc_text))
+                            or not _trunc_text
                         )
                     )
 
@@ -3832,9 +3838,9 @@ def run_conversation(
                     # stripped first (repeated scratchpad lines are not
                     # evidence of a degenerate visible response).
                     _visible_trunc = (
-                        agent._strip_think_blocks(_trunc_content)
-                        if isinstance(_trunc_content, str)
-                        else _trunc_content
+                        agent._strip_think_blocks(_trunc_text)
+                        if _trunc_text
+                        else ""
                     )
                     _repetition_dominated = (
                         not _trunc_has_tool_calls
--- a/agent/message_content.py
+++ b/agent/message_content.py
@@ -4,7 +4,10 @@ from collections.abc import Mapping
 from typing import Any
 
 
-_NON_TEXT_PART_TYPES = {"image", "image_url", "input_image", "audio", "input_audio"}
+_NON_TEXT_PART_TYPES = {
+    "image", "image_url", "input_image", "audio", "input_audio",
+    "tool_use", "tool_call",
+}
 _TEXT_KEYS = ("text", "content", "input_text", "output_text", "summary_text")
 
 
--- a/agent/repetition_guard.py
+++ b/agent/repetition_guard.py
@@ -21,6 +21,9 @@ repeated, code with similar-looking lines) are never blocked.
 from __future__ import annotations
 
 import math
+from typing import Any
+
+from agent.message_content import flatten_message_text
 
 # A fragment must be at least this long before the repetition check runs at
 # all.  Short truncations (a sentence cut mid-word) can trivially contain
@@ -40,7 +43,7 @@ _MIN_REPEAT_COUNT = 5
 _DOMINANCE_RATIO = 0.5
 
 
-def is_repetition_dominated(text: str) -> bool:
+def is_repetition_dominated(text: Any) -> bool:
     """True when ``text`` is dominated by verbatim repeated fragments.
 
     A truncated response is "repetition-dominated" when a single 60+ char
@@ -50,9 +53,12 @@ def is_repetition_dominated(text: str) -> bool:
     continuation nudge would just stitch more repeated text into the final
     response.
 
+    Accepts plain string or multipart list content (which is flattened).
     Returns False for non-string / empty / short inputs (fail-open: never
     blocks a continuation the guard cannot confidently judge).
     """
+    if isinstance(text, list):
+        text = flatten_message_text(text)
     if not isinstance(text, str):
         return False
     n = len(text)
--- a/tests/agent/test_repetition_guard.py
+++ b/tests/agent/test_repetition_guard.py
@@ -63,3 +63,23 @@ class TestRepetitionGuard:
             {"type": "text", "text": "Unique body part detailing verification steps."},
         ]
         assert is_repetition_dominated(parts) is False
+
+    def test_multipart_with_images_and_tool_calls_not_flagged(self):
+        """Non-text parts (images, tool_use) must not be stringified into false repetition signals."""
+        parts = [
+            {"type": "text", "text": "Here is the screenshot requested:"},
+            {"type": "image_url", "image_url": {"url": "data:image/png;base64,AAAA"}},
+            {"type": "tool_use", "id": "call_1", "name": "screenshot", "input": {}},
+        ]
+        assert is_repetition_dominated(parts) is False
+
+    def test_multipart_repeated_table_syntax_below_dominance_not_flagged(self):
+        """Legitimate repeated markdown tables or code syntax below dominance threshold must pass."""
+        table_rows = ["| col1 | col2 | col3 |", "| --- | --- | --- |"] + [
+            f"| data_{i} | value_{i} | result_{i} |" for i in range(100)
+        ]
+        parts = [
+            {"type": "text", "text": "\n".join(table_rows)},
+        ]
+        assert is_repetition_dominated(parts) is False
+
--- a/tests/agent/test_repetition_guard.py
+++ b/tests/agent/test_repetition_guard.py
@@ -48,3 +48,18 @@ class TestRepetitionGuard:
         assert is_repetition_dominated("") is False
         assert is_repetition_dominated(None) is False
         assert is_repetition_dominated(12345) is False
+
+    def test_multipart_list_content_flagged(self):
+        # Multipart list where text parts contain repeated sentences
+        parts = [
+            {"type": "text", "text": _INCIDENT_ECHO * 1000},
+            {"type": "text", "text": _INCIDENT_ECHO * 1000},
+        ]
+        assert is_repetition_dominated(parts) is True
+
+    def test_multipart_list_non_repeating_not_flagged(self):
+        parts = [
+            {"type": "text", "text": "Unique intro part describing architecture."},
+            {"type": "text", "text": "Unique body part detailing verification steps."},
+        ]
+        assert is_repetition_dominated(parts) is False
--- a/tests/run_agent/test_continuation_repetition_guard.py
+++ b/tests/run_agent/test_continuation_repetition_guard.py
@@ -97,3 +97,14 @@ class TestContinuationRepetitionGuard:
 
         assert result["partial"] is True
         assert loop_agent.client.chat.completions.create.call_count == 4
+
+    def test_multipart_list_repetition_aborts(self, loop_agent):
+        echo_list = [{"type": "text", "text": _INCIDENT_ECHO * 2000}]
+        loop_agent.client.chat.completions.create.side_effect = [_stub(echo_list)]
+
+        result = _run(loop_agent, "write me a long report")
+
+        assert result["completed"] is False
+        assert result["partial"] is True
+        assert "Repetition" in (result["final_response"] or "")
+        assert loop_agent.client.chat.completions.create.call_count == 1


alt-glitch:

Related to merged #86698 and #86581. This is a follow-up for multipart list content, which the original continuation guard did not flatten; it is not a duplicate of the merged string-content repair.

liyangbing:
<h3>Review</h3>

<p>Flattening multipart content at the guard boundary closes the type error and restores protection for structured model responses. The important semantic constraint is to detect repetition in visible assistant text without treating every non-text block or repeated code fragment as a degenerate continuation.</p>

<h3>Suggested contract</h3>

<ul>
<li>Make flatten_message_text return a clearly scoped visible-text projection; do not stringify images, tool arguments, citations, or opaque metadata into the repetition signal.</li>
<li>Preserve the distinction between an empty visible projection, a non-text-only response, and a text projection that is genuinely repetition-dominated. None should trigger a false continuation abort solely because the list is non-empty.</li>
<li>Keep the guard decision tied to the current attempt and finish reason, with an explicit bounded continuation budget and an operator-visible reason; do not let a successful long response or a legitimate repeated code/table pattern become an infinite-loop diagnosis.</li>
<li>Ensure think-tag extraction, whitespace normalization, Unicode normalization, and multipart ordering are deterministic and do not mutate the content later persisted or shown to the user.</li>
<li>Keep the fallback fail-safe: if flattening encounters an unknown block, record a bounded diagnostic and avoid the old TypeError, while retaining enough evidence to distinguish an unguarded response from a deliberate abort.</li>
</ul>

<h3>Regression matrix</h3>

<ul>
<li>plain string, one text block, multiple text blocks, text plus image, text plus tool-call, and reasoning-only multipart content</li>
<li>repeated text split across block boundaries, repeated whitespace, Unicode variants, and think-tag-wrapped content</li>
<li>legitimate repeated code, tables, logs, quoted text, and long non-repeating output</li>
<li>finish_reason length, stop, tool-call completion, cancellation, provider error, and malformed/unknown blocks</li>
<li>single continuation, repeated continuation attempts, and budget exhaustion with no channel flood</li>
<li>conversation persistence and replay proving the guard projection does not rewrite the original assistant content</li>
<li>Anthropic, Gemini, Bedrock, OpenAI Responses, and any adapter that emits nested content lists</li>
</ul>

<p>The handbook's <a href="https://sandbaseai.github.io/deepseek-harness-handbook/degenerate-model-output.html">degenerate-output guide</a> follows the same separation between visible output, structured tool content, continuation budgets, and operator evidence.</p>

PR #90889, opened 2026-08-20
fix(delegate): defer timed-out child teardown

ComicBit:
## What does this PR do?

Prevents timed-out delegated children from being closed while their conversation thread is still unwinding.

`_run_single_child()` waits on a child `Future`. On timeout it requests a cooperative hard interrupt, but the function's outer `finally` immediately called `child.close()` even when the `Future` was still running. `AIAgent.close()` closes the child's owned `SessionDB`; meanwhile `run_conversation()` can still be executing its final activity-label write on the worker thread. That cross-thread close/use race matches the observed gateway segfault stack.

When a timed-out child `Future` is still running, this change transfers close ownership to a `Future` done-callback. Normal completion and already-settled failure paths keep their existing synchronous close behavior.

## Related Issue

No existing issue found. This was reproduced on current `main` after a gateway crash involving two simultaneous delegated-child timeouts.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- Defer `child.close()` until the timed-out worker `Future` has fully settled.
- Preserve synchronous teardown for normal and already-settled paths.
- Add a deterministic regression test that holds the fake child inside its unwind phase and asserts teardown cannot race it.

## How to Test

1. Run `scripts/run_tests.sh tests/tools/test_delegate_timeout_cleanup.py -q`.
2. Temporarily restore the old unconditional `child.close()` path; the regression test fails with `timed-out child.close() ran before its conversation thread unwound`.
3. Run the delegation/subagent family:
   `scripts/run_tests.sh tests/tools/test_delegate*.py tests/tools/test_subagent*.py tests/agent/test_subagent*.py tests/test_delegate_cascade_49148.py tests/cli/test_cli_delegate_background_notice.py tests/cli/test_cli_interrupt_subagent.py -q`

Local results:
- Regression test with fix: `1 passed`
- Sabotage run against old behavior: `1 failed` at the intended assertion
- Delegation/subagent suite: `255 passed, 0 failed`
- `ruff 0.15.10`: clean
- `git diff --check`: clean

### Code

[...]

PR #98348, opened 2026-08-30
fix(delegation): keep timed-out children from crashing session owners

### LINKED ISSUE #98332: Managed Python 3.11.15 SIGSEGV in sqlite3 C extension (_PyDict_GetItem_KnownHash via bounded_lru_cache_wrapper) at child 600s delegation timeout kills the owner process; batch reported as 'owner exited'
## Description

The managed CPython 3.11.15 interpreter (uv-managed, `com.nousresearch.hermes.managed-python`, ad-hoc signed) crashes with a native SIGSEGV inside the `sqlite3` C extension, on a functools `lru_cache` wrapper dict lookup. The crash kills the whole process that hosts the active agent session (the async-delegation "owner"), so in-flight delegation batches are later classified as outcome-unknown with the fixed message `Delegation owner exited before recording a terminal result; outcome unknown.` (emitted by `recover_abandoned_delegations()` in `tools/async_delegation.py:380` after the dead owner pid is detected).

## Environment

- Hermes Agent: v0.20.6 (2026.8.27), upstream `4209d371`, install method: git (`hermes update` workflow)
- OS: macOS 26.6.2 (25G83), Apple Silicon (arm64)
- Python: managed runtime 3.11.15 (`3.11.15 (main, Jul 18 2026, 16:57:24) [Clang 22.1.3]`, platform macosx-11.0-arm64) at `.hermes-runtime/python/generation-1785114051-80376-69c46964/cpython-3.11.15-macos-aarch64-none/`, venv at `~/.hermes/hermes-agent/venv` (pyvenv.cfg `home` points at that runtime); runtime generation dated Jul 27 2026, venv python re-anchored/re-signed 2026-08-29 11:04 (mtime)
- Gateway under launchd (`ai.hermes.gateway.plist`, `KeepAlive`)
- `delegation.child_timeout_seconds: 600`

## Symptom

Two native SIGSEGVs of the python3 process (crash reports in `~/Library/Logs/DiagnosticReports/`), each killing a live async-delegation owner:

| crash report (.ips) | faulting pid | matching `async_delegations` row |
|---|---|---|
| python3-2026-08-30-033831.ips | 2274 | `deleg_8eef4acc` (owner_pid=2274) → state `unknown` |
| python3-2026-08-30-040030.ips | 10992 | `deleg_8fd4201a` (owner_pid=10992) → state `unknown` |

`codeSigningID: com.nousresearch.hermes.managed-python`, `codeSigningFlags: 0x22000201` (adhoc, CS_KILL set).

Both faulting threads are identical, non-main threads (a worker thread running a SQLite query), exception `EXC_BAD_ACCESS (SIGSEGV) KERN_INVALID_ADDRESS at 0x0000000000000008` (null-deref + field offset):

```
_pysqlite_query_execute+268
  PyObject_Vectorcall+396
  bounded_lru_cache_wrapper+292
  _PyDict_GetItem_KnownHash+32     <-- faulting instruction
```

The faulting dict key lookup is on an lru-cached function that is invoked during `pysqlite_connection.execute(...)`. The same binary's symbol table shows `bounded_lru_cache_wrapper` / `uncached_lru_cache_wrapper` / `infinite_lru_cache_wrapper` (the 3.12+ naming in a 3.11 build).

## Timing pattern (strong)

Both crashes occur within ~1s of a child hitting the 600s hard timeout:

- 03:28:25 delegation `deleg_8eef4acc` dispatched → 03:38:31 crash (606s) → batch marked unknown 03:38:57
- 03:50:29 delegation `deleg_8fd4201a` dispatched → child log shows `final | status=timeout duration=600.01s` at 04:00:29 → crash 04:00:30

Counter-evidence on the same owners:
- owner pid 2274 completed 3 earlier batches (deleg_74c72ab2 / d114ee9b / 21d2e54c, all 03:04–03:40) without crashing
- owner pid 10992 completed deleg_c27b08f7 (03:40:01) without crashing
- after 04:00, two re-dispatched batches under owner pid 47666 (child durations 353s / 325s, both < 600s) completed fine
- no further python crash reports on this machine after 04:00 on 08-30

## Correlation with macOS upgrade (unverified, flagged as such)

- macOS 26.6.1 → 26.6.2 installed 2026-08-29 16:17 (softwareupdate log); reboot 2026-08-30 02:49
- The managed python binary at `venv/bin/python` has mtime 2026-08-29 11:04 (uv-managed rotation/TCC re-anchor per `hermes_cli/macos_tcc_anchor.py`); an adjacent `python-2026-08-29-110509.ips` shows a one-off `SIGKILL (Code Signature Invalid)` at 11:05 (different failure mode, self-resolved)
- No SIGSEGV crash reports exist on this machine before 2026-08-30 03:38
- I cannot rule out that the 26.6.2 kernel change merely exposed a latent CPython thread-safety bug rather than causing it

## Questions for maintainers

1. Does the managed 3.11.15 build backport the 3.12 `bounded_lru_cache` rework onto 3.11 sources, and is that combination known to be thread-safe under concurrent sqlite3 use (multiple threads executing queries, one thread timing out/killing a child)?
2. Is there a supported way to re-provision/rotate the managed python runtime (e.g. after such a crash or after a macOS upgrade) without a full reinstall?
3. Should `recover_abandoned_delegations()` distinguish "owner killed by signal (check crash report / exit status)" from "owner still alive but slow", so the surfaced message points at the real cause?

## Workaround in place

Keep child work under 600s (tightened sub-agent tool-call budgets so completion is ~300–360s). This avoids the crash window but is not a fix.

## Artifacts

The two .ips crash reports can be shared on request (they contain only system/library symbols, no user data).

fangliquanflq:
## What does this PR do?

A configured subagent wall-clock timeout can currently close the child agent while its worker thread is still unwinding. That closes the child's SQLite-backed SessionDB under an active query or flush, which can lose delegation state and, on the reported managed macOS runtime, coincides with a native sqlite3 SIGSEGV that kills the session owner. This change transfers teardown to the worker completion boundary and makes abandoned-owner recovery report only process identity evidence it can actually prove.

### Symptom

With `delegation.child_timeout_seconds` configured, a child that does not stop immediately after interruption reaches the timeout path while its worker remains inside `run_conversation`. The owner then calls `child.close()` from the waiting thread. The reported environment observed two native sqlite3 SIGSEGV crashes within about one second of the 600 second timeout, followed by recovery as a generic `owner exited` outcome.

### Impact

The owner process can lose the active session and every in-flight delegation in the batch. Recovery preserves an unknown result, but the existing generic message does not distinguish a missing owner process from PID reuse or explain that signal and exit status are not observable from the recovering process.

### Bug Cause

**Trigger:** `tools/delegate_tool.py:_run_single_child` when `_child_future.result(timeout=child_timeout)` times out and interruption does not immediately stop the worker.

**Causal chain:**
1. The timeout branch requests interruption but cannot synchronously stop a daemon worker already inside a provider, tool, or SQLite-backed session operation.
2. `_run_single_child` returns through `finally` and calls `child.close()` even though `_child_future.done()` is false.
3. `AIAgent.close()` closes the child's dedicated SessionDB and other resources while the worker can still use them, creating a teardown/use race. On the reported managed runtime, the race coincides with a native sqlite3 crash and kills the owner.

**Why it is wrong:** Resource ownership crosses the worker lifetime boundary. A timeout abandons the wait, not the still-running worker, so the waiting thread cannot safely close worker-owned resources.

**Working sibling / contrast:** Normal completion and non-timeout exceptions have a completed future before final teardown, so closing on the waiting thread is safe. The fix changes only the timed-out, still-running future path.

**Ruled out:** The recovery code does not kill owners and does not classify a live owner with a matching PID/start stamp as abandoned. The regression reproduces the premature close directly without assuming the managed Python build is the sole cause of the native crash.

### Fix

- When a timeout leaves the child future running, register a done callback that retires the child exactly once after the worker exits. Until then, keep its live registry entries, parent interrupt-propagation membership, credential lease, and owned resources intact.
- Classify recovery evidence as `owner_process_missing`, `owner_pid_reused`, or `owner_identity_missing`.
- Keep live owners when the start-time probe is indeterminate instead of falsely recovering a potentially slow owner.
- Persist structured `owner_exit` evidence with explicit null `exit_code` and `signal` fields. The recovery process cannot obtain another process's wait status, so the user-facing error points operators to supervisor or OS crash evidence rather than inventing an exit cause.

## Related Issue

Addresses #98332

## Type of Change

## Changes Made

- `tools/delegate_tool.py` - transfer the complete timed-out child lifetime retirement to the child future completion boundary.
- `tools/async_delegation.py` - add fail-closed owner identity classification and structured recovery evidence.
- `tests/tools/test_delegate_subagent_timeout_diagnostic.py` - reproduce the premature close and verify teardown occurs exactly once after worker exit.
- `tests/tools/test_delegate_timeout_cleanup.py` - preserve the prior delayed-unwind regression from #90889 with its original authorship.
- `tests/tools/test_async_delegation.py` - cover live, missing, reused, and untracked owner classifications.

## How to Test

1. Configure a short child timeout and run a child that ignores interruption until a test gate is released.
2. Verify the timeout result returns while `child.close()` remains uncalled, release the worker, and verify close runs exactly once.
3. Run the relevant delegation and durable-ledger suites:

```bash
scripts/run_tests.sh tests/tools/test_delegate_timeout_cleanup.py tests/tools/test_delegate_subagent_timeout_diagnostic.py tests/tools/test_delegate.py tests/tools/test_async_delegation_fd_leak.py tests/tools/test_async_delegation.py -q
```

Result: 111 passed, 1 macOS-only test skipped on Windows 11.

### Code

- [x] I've run the repository test entry on the relevant tests and all tests pass

- [x] Documentation update is N/A; recovery behavior is described by code docstrings and result fields
- [x] I've considered cross-platform impact; the lifetime fix is platform-independent and existing macOS-only SQLite barriers remain covered on the macOS CI lane
- [x] Tool description/schema update is N/A; no model tool schema changed

## Screenshots / Logs

The targeted test suite passed with 111 tests. The exact managed macOS runtime reproduction remains delegated to the review environment because the native crash artifact is macOS-specific, so this PR addresses rather than closes #98332.

### DIFF
--- a/contributors/emails/ileocorp@gmail.com
+++ b/contributors/emails/ileocorp@gmail.com
@@ -1,0 +1,2 @@
+ComicBit
+# PR #98348 preserves commit 057000ed73c from #90889
--- a/tests/tools/test_async_delegation.py
+++ b/tests/tools/test_async_delegation.py
@@ -153,6 +153,43 @@ def test_active_for_session_counts_every_live_delegation_state():
     assert ad.active_for_session("") == 0
 
 
+def test_owner_process_classification_preserves_indeterminate_live_owner():
+    """A failed start-time probe must not turn a slow live owner into an orphan."""
+    assert ad._classify_owner_process(
+        4321,
+        100,
+        pid_exists=lambda _pid: True,
+        process_start_time=lambda _pid: None,
+    ) is None
+
+
+@pytest.mark.parametrize(
+    ("pid", "started_at", "exists", "observed_start", "reason"),
+    [
+        (None, None, False, None, "owner_identity_missing"),
+        (4321, 100, False, None, "owner_process_missing"),
+        (4321, 100, True, 200, "owner_pid_reused"),
+    ],
+)
+def test_owner_process_classification_reports_only_observable_evidence(
+    pid, started_at, exists, observed_start, reason
+):
+    evidence = ad._classify_owner_process(
+        pid,
+        started_at,
+        pid_exists=lambda _pid: exists,
+        process_start_time=lambda _pid: observed_start,
+    )
+
+    assert evidence is not None
+    assert evidence["reason"] == reason
+    assert evidence["exit_code"] is None
+    assert evidence["signal"] is None
+    assert "check the process supervisor or OS crash report" in ad._owner_exit_error(
+        evidence
+    )
+
+
 def test_dispatch_returns_immediately_without_blocking():
     gate = threading.Event()
 
--- a/tests/tools/test_delegate_subagent_timeout_diagnostic.py
+++ b/tests/tools/test_delegate_subagent_timeout_diagnostic.py
@@ -211,6 +211,23 @@ class TestRunSingleChildTimeoutDump:
         assert "Diagnostic:" in result["error"]
         assert str(dump_path) in result["error"]
 
+    def test_timeout_defers_child_close_until_worker_exits(self, hermes_home, monkeypatch):
+        """A detached worker must keep its persistence handle until it stops."""
+        child = _StubChild(api_call_count=1, hang_seconds=10.0)
+        child.interrupt = lambda *args, **kwargs: None
+        child.close = MagicMock()
+
+        result = self._invoke_with_short_timeout(child, monkeypatch)
+
+        assert result["status"] == "timeout"
+        assert not child.close.called
+
+        child._hang.set()
+        deadline = time.monotonic() + 2.0
+        while not child.close.called and time.monotonic() < deadline:
+            time.sleep(0.01)
+        assert child.close.call_count == 1
+
 
     # ── explicit timeout metadata (#51690, salvaged from PR #60378) ────
 
--- a/tests/tools/test_delegate_subagent_timeout_diagnostic.py
+++ b/tests/tools/test_delegate_subagent_timeout_diagnostic.py
@@ -212,21 +212,29 @@ class TestRunSingleChildTimeoutDump:
         assert str(dump_path) in result["error"]
 
     def test_timeout_defers_child_close_until_worker_exits(self, hermes_home, monkeypatch):
-        """A detached worker must keep its persistence handle until it stops."""
+        """A detached worker retains its persistence handle and credential lease."""
         child = _StubChild(api_call_count=1, hang_seconds=10.0)
         child.interrupt = lambda *args, **kwargs: None
         child.close = MagicMock()
+        child._credential_pool = MagicMock()
+        child._credential_pool.acquire_lease.return_value = "cred-a"
+        child._credential_pool.current.return_value = MagicMock(id="cred-a")
 
         result = self._invoke_with_short_timeout(child, monkeypatch)
 
         assert result["status"] == "timeout"
         assert not child.close.called
+        assert not child._credential_pool.release_lease.called
 
         child._hang.set()
         deadline = time.monotonic() + 2.0
-        while not child.close.called and time.monotonic() < deadline:
+        while (
+            not child._credential_pool.release_lease.called
+            and time.monotonic() < deadline
+        ):
             time.sleep(0.01)
         assert child.close.call_count == 1
+        child._credential_pool.release_lease.assert_called_once_with("cred-a")
 
 
     # ── explicit timeout metadata (#51690, salvaged from PR #60378) ────
--- a/tests/tools/test_delegate_timeout_cleanup.py
+++ b/tests/tools/test_delegate_timeout_cleanup.py
@@ -1,0 +1,90 @@
+"""Regression coverage for timed-out delegation teardown."""
+
+from __future__ import annotations
+
+import threading
+from types import SimpleNamespace
+
+from tools import delegate_tool
+
+
+class _SlowUnwindingChild:
+    def __init__(self) -> None:
+        self.tool_progress_callback = None
+        self._credential_pool = None
+        self._delegate_saved_tool_names = []
+        self._delegate_role = "leaf"
+        self._delegate_depth = 1
+        self._subagent_id = None
+        self.model = "test-model"
+        self.session_prompt_tokens = 0
+        self.session_completion_tokens = 0
+        self.session_estimated_cost_usd = 0.0
+        self.session_cost_status = "unknown"
+        self.started = threading.Event()
+        self.interrupted = threading.Event()
+        self.unwinding = threading.Event()
+        self.allow_finish = threading.Event()
+        self.finished = threading.Event()
+        self.closed = threading.Event()
+        self.close_while_running = False
+
+    def run_conversation(self, **_kwargs):
+        self.started.set()
+        assert self.interrupted.wait(timeout=1)
+        # Model the real child turn's finally path: it still performs session
+        # activity/SQLite cleanup after the parent requests interruption.
+        self.unwinding.set()
+        assert self.allow_finish.wait(timeout=2)
+        self.finished.set()
+        return {
+            "final_response": "",
+            "completed": False,
+            "interrupted": True,
+            "api_calls": 1,
+            "messages": [],
+        }
+
+    def hard_interrupt(self, _reason=None):
+        self.interrupted.set()
+
+    def get_activity_summary(self):
+        return {"api_call_count": 1}
+
+    def close(self):
+        if not self.finished.is_set():
+            self.close_while_running = True
+        self.closed.set()
+
+
+def test_timeout_does_not_close_child_while_worker_is_unwinding(monkeypatch):
+    child = _SlowUnwindingChild()
+    parent = SimpleNamespace(
+        session_id="parent-timeout-test",
+        _current_task_id=None,
+        _active_children=[child],
+        _active_children_lock=threading.Lock(),
+    )
+    monkeypatch.setattr(delegate_tool, "_get_child_timeout", lambda: 0.5)
+    monkeypatch.setattr(delegate_tool, "_get_worktree_isolation", lambda: False)
+
+    result = delegate_tool._run_single_child(
+        task_index=0,
+        goal="exercise timeout teardown",
+        child=child,
+        parent_agent=parent,
+    )
+
+    assert result["status"] == "timeout"
+    assert child.unwinding.wait(timeout=1)
+    try:
+        assert not child.closed.is_set(), (
+            "timed-out child.close() ran before its conversation thread unwound"
+        )
+    finally:
+        child.allow_finish.set()
+    assert child.finished.wait(timeout=1)
+    assert child.closed.wait(timeout=1)
+    assert not child.close_while_running, (
+        "timed-out child.close() raced its still-running conversation thread"
+    )
--- a/tools/async_delegation.py
+++ b/tools/async_delegation.py
@@ -340,8 +340,71 @@ def _note_delivery_attempt(delegation_id: str) -> None:
         )
 
 
+def _classify_owner_process(
+    pid: Optional[int],
+    started_at: Optional[int],
+    *,
+    pid_exists: Callable[[int], bool],
+    process_start_time: Callable[[int], Optional[int]],
+) -> Optional[Dict[str, Any]]:
+    """Return durable owner-exit evidence, or ``None`` while owner may be live.
+
+    Recovery runs in a different process, so it cannot obtain the dead
+    process's wait status.  Report only process-table facts and leave signal /
+    exit code explicitly unknown rather than guessing that every disappearance
+    was a graceful exit.  A live PID whose start time cannot be inspected is
+    treated as indeterminate/live; recovering it would steal work from a slow
+    owner merely because the identity probe failed.
+    """
+    evidence: Dict[str, Any] = {
+        "reason": "owner_identity_missing",
+        "pid": int(pid) if pid else None,
+        "started_at": int(started_at) if started_at is not None else None,
+        "observed_started_at": None,
+        "exit_code": None,
+        "signal": None,
+        "evidence": "process_table",
+    }
+    if not pid:
+        return evidence
+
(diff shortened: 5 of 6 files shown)


alt-glitch:

Related to #90889's open timeout-teardown repair and #98332's crash report. This implementation uses the same safe Future-completion teardown boundary while also adding structured owner-process recovery evidence; please consolidate the competing fixes before merge.

andrexibiza (commented):
Reviewed exact head `13b4a069abfb77dcb7df567c063062fd5f30d8cf` against its exact parent/current `main@26350357d76e4508c8df9304a3374bdc5a6f6220`. The object is fully green now: CI `33292962467`, Docker `33292962117`, and Nix `33292962122` all completed successfully, including attribution. I also read the behavior-critical lifetime/recovery paths, the #98332 incident record, existing discussion, and the competing/adjacent delegation work.

There is real good work here. The core observation is correct: a timeout expires the owner's wait; it does **not** prove the worker has stopped. Moving `child.close()` to the Future completion boundary is the right direction, and the recovery classifier is appropriately conservative: it distinguishes missing PID vs PID reuse, treats an indeterminate start-time probe as potentially live, and explicitly refuses to manufacture signal/exit status that this process cannot observe.

### P1 — the lifetime transfer is incomplete: the still-running child loses other ownership before its Future settles

The new timeout path in `tools/delegate_tool.py` sets `_defer_child_close_to_worker = True` and attaches `_close_detached_child()` to the still-running Future. But the outer finalizer still retires several **other** pieces of child ownership immediately, before that Future is done:

- it calls `_unregister_subagent(_subagent_id, agent=child)`, removing the still-running child from the live control/observability registry;
- it calls `child_pool.release_lease(leased_cred_id)`, returning the credential lease while that same worker may still be inside a provider/tool/unwind path using the credential it acquired at start;
- it removes `child` from `parent_agent._active_children`;
- only `child.close()` is protected by `_defer_child_close_to_worker` in the new hunk around head lines ~3395+.

That is the other side of the exact race this PR is fixing. A timed-out worker can now continue executing after the parent has made it invisible to the parent/control plane and made its leased credential available for reuse. With the PR's own adversarial shape — interrupt requested but the worker does not immediately stop — the resource/liveness generation is still current until the Future actually settles.

Please make worker completion the **single retirement boundary** for the whole child lifetime, not just `close()`: keep the live ownership/lease until the Future is done, then unregister, release the lease, remove the active-child reference, and close exactly once from one bounded retirement routine. It is fine for the public result to return `timeout` immediately; the internal child can be marked `timed_out`/`unwinding`, but its ownership must remain represented until it actually stops.

The regression should hold a non-cooperative worker after timeout and prove, before releasing it, that (1) the child has not been closed, (2) its credential lease has not been released/reacquired, and (3) its live ownership has not been retired; after release, all of those retire exactly once. That closes the *class*, not only the SQLite handle symptom.

### Landing blocker — converge the existing timeout-teardown implementation and preserve its author

Open #90889 by `ComicBit` already implements the same Future-done teardown boundary in the same `tools/delegate_tool.py` path. Its patch predates this PR and includes a purpose-built `_SlowUnwindingChild` regression that asserts the child is not closed while `run_conversation()` is still unwinding. #98348 is not merely a duplicate because `fangliquanflq` adds the structured owner-process recovery work in `tools/async_delegation.py`; that is a useful, distinct contribution. But the teardown mechanism has two competing owners right now.

Please converge them rather than landing parallel copies. The clean shape is: preserve `ComicBit`'s prior timeout-teardown contribution/provenance from #90889, then layer `fangliquanflq`'s owner-recovery evidence and the full-lifetime retirement fix above it. If #98348 becomes the superseding landing object, make that origin chain explicit rather than silently replacing #90889.

### Merge-order interlock — #98349 is complementary but collides in the durable recovery owner

New #98349 by `ericmaddox` directly edits both `tools/async_delegation.py` and `tests/tools/test_async_delegation.py` to harden async finalization/serialization. It is a different defect class, not duplicate work, but it is a same-owner/same-file collision with this PR's recovery changes. Whichever lands second needs explicit current-main composition and fresh exact-head receipts; do not let either overwrite the other's terminal-state/recovery semantics.

### Hard structural gate — this adds more lifecycle authority to an existing >2K godfile

The modified `tools/delegate_tool.py` hunk is already at ~line 3395, so this file is unambiguously above the repository's 2K ceiling. This PR adds another resource-ownership transition directly inside that monolith. The full-lifetime retirement mechanism above is a natural bounded seam: extract the child lifetime/retirement owner instead of adding another authority branch to the godfile. That also gives #90889/#98348 a single convergence surface instead of another in-place collision.

### Closure topology — #98332 still needs exact crash closure evidence

The deterministic application race is real and this PR can close that race. But #98332 is specifically the managed macOS Python/sqlite native SIGSEGV incident, and this PR's own body correctly says the exact reported managed-runtime reproduction has not been run here. The incident also explicitly leaves the managed-runtime contribution unverified. I would not auto-close the entire crash issue on a plausible causal repair alone: use `Addresses #98332` until the reported runtime/stress witness shows the timeout no longer produces the crash, or split the issue so this PR closes the proven teardown race while the runtime-specific crash investigation remains open.

So: the direction is strong and the recovery evidence is substantially better than the current generic `owner exited` message. The remaining work is to make the worker/Future boundary own **all** of the child's lifetime, converge the already-authored teardown fix with credit intact, compose the adjacent durable-finalization work, and keep the native-crash closure claim at exactly the strength the evidence supports. Once those are true, this becomes a much stronger closure for the timeout/session-owner failure class.

liyangbing:
<h3>Review</h3>

<p>Moving child teardown to the worker completion boundary fixes the central ownership race, and the explicit owner-identity classifications are safer than inferring a signal or exit code the recovering process cannot observe.</p>

<h3>Suggested contract</h3>

<ul>
<li>Keep the timeout result distinct from worker termination: a timed-out wait means the caller stopped waiting, not that the child or its resources are already closed.</li>
<li>Register the completion callback before any path can observe a completed future, and guard the close operation with an atomic once-state that covers callback, normal completion, cancellation, and owner shutdown.</li>
<li>Make the callback's failure observable without allowing an exception in cleanup to strand the delegation in finalizing; record a bounded cleanup outcome and preserve the original timeout outcome.</li>
<li>Keep worker-owned SessionDB, provider clients, file descriptors, and temporary resources owned by the worker lifetime. The waiting thread must not reuse or mutate them after abandoning the wait.</li>
<li>Fence the child by delegation id and owner generation so a late callback cannot close a replacement child or write a terminal result over a newer attempt.</li>
<li>Keep recovery conservative when the process-identity probe is indeterminate, and distinguish live-but-slow, missing, reused, and untracked identities in both durable state and operator output.</li>
</ul>

<h3>Regression matrix</h3>

<ul>
<li>future completes before timeout, during timeout handling, and after the callback is registered</li>
<li>interruption succeeds immediately, is delayed, or is ignored by provider/tool/SQLite work</li>
<li>callback close succeeds, raises, or races with owner shutdown</li>
<li>single and batch delegation with late child completion, duplicate callbacks, and owner restart</li>
<li>live owner, missing owner, PID reuse, missing start stamp, and indeterminate process probe</li>
<li>late completion after a retry or replacement delegation, proving generation fencing</li>
<li>active-count, durable-ledger, file-descriptor, and SessionDB cleanup after success, timeout, cancellation, and partial failure</li>
</ul>

<p>The handbook's <a href="https://sandbaseai.github.io/deepseek-harness-handbook/stuck-tool-cancellation.html">cancellation guide</a> follows the same distinction between cancellation admission, live work, ownership, and terminal recovery evidence.</p>

fangliquanflq:
Thanks for the detailed review. I investigated each lifecycle and recovery claim against the exact PR head and current base, then pushed `00a6a0c6c8a`.

- **Full child-lifetime retirement — valid and fixed.** A timed-out, still-running worker now retains its active-subagent registration, parent interrupt-propagation membership, credential lease, and owned agent resources. One lock-guarded retirement callback releases all four only after the Future settles. Each cleanup step is isolated, so a close failure cannot strand the lease or overwrite the original timeout result.
- **Prior #90889 implementation/provenance — valid and preserved.** I merged ComicBit's original commit `057000ed73c` into this branch, retaining Leandro Piccione's authorship and delayed-unwind regression, then composed the broader lifetime fix on top.
- **#98349 — informational.** It addresses a separate durable serialization/finalization defect but overlaps `async_delegation.py`; whichever change lands second will need current-main composition. I did not mix that separate fix into this PR.
- **Godfile extraction — no current defect or repository hard gate.** `delegate_tool.py` is large, but neither `AGENTS.md` nor `CONTRIBUTING.md` defines a 2K-line ceiling. Extracting a lifecycle owner may be a useful focused follow-up; doing it here would broaden this concurrency fix and increase collision risk.
- **Native crash closure strength — valid and corrected.** The PR now says `Addresses #98332`, not `Closes #98332`, because the deterministic teardown race is covered but the exact managed macOS Python/sqlite crash witness has not been rerun.

For the suggested contract and matrix: timeout remains distinct from worker termination; callback registration is race-safe through `Future.add_done_callback`; retirement is exact-once; cleanup exceptions are logged without replacing the timeout result; live/missing/reused/untracked/indeterminate owner identities remain conservatively classified. A new regression holds a non-cooperative child after timeout and proves all live ownership remains before release and retires exactly once afterward; another proves a failing `close()` still releases the lease. A generation fence is not added because the callback captures the exact child object, identity-checks unregister, and performs no terminal-result write or replacement lookup.

Validation:

```text
scripts/run_tests.sh tests/tools/test_delegate_timeout_cleanup.py tests/tools/test_delegate_subagent_timeout_diagnostic.py tests/tools/test_delegate.py tests/tools/test_async_delegation_fd_leak.py tests/tools/test_async_delegation.py -q
111 passed, 1 skipped (macOS-only), 0 failed
```

An independent scan-only review of committed tip `00a6a0c6c8a` found no in-scope issues (`verdict: perfect`).

PR #71821, opened 2026-07-26
fix(terminal): surface scrubbed provider credentials in tool results

Bartok9:
## Summary
Closes #71788. When Hermes scrubs provider credentials from terminal/`execute_code` child envs, return a `credential_scrub_note` (and output footer) naming the removed vars so "KEY is not set" is not mistaken for misconfiguration. Scrub policy unchanged.

## Verification
`python3 -m pytest tests/tools/test_env_passthrough.py -q`

PR #98354, opened 2026-08-30
fix(terminal): surface scrubbed provider credentials in tool results (#71788) (salvage of #71821)

### LINKED ISSUE #71788: DX: surface when terminal scrub removes a provider credential (silent 'not set')
## Context

Hermes intentionally scrubs Hermes-managed provider credentials (e.g. `DASHSCOPE_API_KEY` from the alibaba/Qwen Cloud provider registry) from `terminal` / `execute_code` child environments. That matches GHSA-rhgp-j443-p4rf and `tools/env_passthrough.py` (passthrough cannot override the blocklist). **This issue is not asking to weaken that scrub.**

## Problem (DX)

The scrub is **silent**. From the agent / model point of view the variable simply looks unset:

```text
KEYLEN=0; WORKSPACE=…; REGION=beijing
{"status":"error","reason":"DASHSCOPE_API_KEY is not set"}
```

Meanwhile the **gateway process** still has the variable. Operators and models then waste turns hunting for the key (`.env`, keychain, `ps`, inventing workarounds) instead of understanding that policy removed it from the sandbox.

We hit this in production with a skill that uses `DASHSCOPE_API_KEY` as a **tool** credential for Qwen Image generation, while the same name is also a Hermes **inference-provider** credential. Gateway env was correct; only the scrubbed terminal child failed.

## Proposal (non-breaking)

Any one of:

1. **Terminal / execute_code result annotation** when a variable present in the parent env was stripped by `_HERMES_PROVIDER_ENV_BLOCKLIST`, e.g. append a short stderr/footer:
   > `note: DASHSCOPE_API_KEY was removed from this sandbox by Hermes credential scrub (provider blocklist). It is not missing from the gateway process.`
2. **`hermes doctor` / hooks doctor check** that lists scrubbed provider env vars and clarifies they will not appear in terminal children.
3. Optional docs note under terminal env / SECURITY that tool scripts sharing a provider env var name must load credentials from a file / secret store inside the child, not expect inheritance.

No change to the scrub policy itself.

## Why it matters

Silent scrub looks identical to misconfiguration. Models escalate to unsafe recovery behavior (searching for keys, asking users for alternate sources) instead of stopping with a clear policy explanation.

## Environment

- Hermes ~v0.19 gateway + local terminal backend
- Observed with `DASHSCOPE_API_KEY` (registered in `PROVIDER_REGISTRY` for alibaba / alibaba-coding-plan)

Bartok9:
## Summary
Salvage of #71821 onto current `main`. Closes #71788.

When Hermes scrubs provider credentials from terminal/`execute_code` child envs, return a `credential_scrub_note` naming the removed vars so "KEY is not set" is not mistaken for misconfiguration. Scrub policy unchanged (GHSA-rhgp-j443-p4rf).

## Rebase notes
Main now builds local execute_code env via `_build_child_env`; the salvage computes the note from that actual child env (instead of the older inline `_scrub_child_env` path).

Original author: @Bartok9 (#71821).

## Verification
`python3 -m pytest tests/tools/test_env_passthrough.py -q` → 31 passed

## Supersedes
Closes the stale conflicting PR #71821 in favor of this fresh base.

### DIFF
--- a/tests/tools/test_env_passthrough.py
+++ b/tests/tools/test_env_passthrough.py
@@ -411,3 +411,153 @@ class TestTerminalIntegration:
         assert "OPENAI_API_KEY" not in child_env
         assert "ANTHROPIC_API_KEY" not in child_env
         assert child_env["PATH"] == "/usr/bin"
+
+
+class TestScrubbedProviderCredentialNotes:
+    """#71788: surface when provider credentials are scrubbed (DX only)."""
+
+    def test_list_and_format_single(self):
+        from tools.env_passthrough import (
+            format_scrubbed_provider_env_note,
+            list_scrubbed_provider_credentials,
+        )
+
+        names = list_scrubbed_provider_credentials(
+            {"DASHSCOPE_API_KEY": "secret", "PATH": "/usr/bin"},
+            {"PATH": "/usr/bin"},
+        )
+        assert "DASHSCOPE_API_KEY" in names
+        note = format_scrubbed_provider_env_note(names)
+        assert "DASHSCOPE_API_KEY" in note
+        assert "credential scrub" in note
+        assert "not missing from the gateway" in note
+
+    def test_empty_parent_value_not_reported(self):
+        from tools.env_passthrough import list_scrubbed_provider_credentials
+
+        names = list_scrubbed_provider_credentials(
+            {"DASHSCOPE_API_KEY": "", "PATH": "/usr/bin"},
+            {"PATH": "/usr/bin"},
+        )
+        assert "DASHSCOPE_API_KEY" not in names
+
+    def test_present_in_child_not_reported(self):
+        from tools.env_passthrough import list_scrubbed_provider_credentials
+
+        names = list_scrubbed_provider_credentials(
+            {"OPENAI_API_KEY": "sk-x", "PATH": "/usr/bin"},
+            {"OPENAI_API_KEY": "sk-x", "PATH": "/usr/bin"},
+        )
+        assert "OPENAI_API_KEY" not in names
+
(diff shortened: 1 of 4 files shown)


Enough1122:

**Overall:** DX-only surfacing of scrubbed provider credentials without changing scrub policy.

**What it does**
- `tools/env_passthrough.py` adds `list_scrubbed_provider_credentials` (parent vs child diff over `_HERMES_PROVIDER_ENV_BLOCKLIST`, ignoring empty parent values) and `format_scrubbed_provider_env_note` (one-line note, singular/plural, truncated at 8 names).
- `tools/code_execution_tool.py` computes note from actual `child_env` for local path, attaches as `credential_scrub_note` dedicated field (never mutating `output`); remote path explicitly leaves note empty since remote env not modeled.
- `tools/terminal_tool.py` for local env derives child via `_make_run_env(self.env)` and both attaches field and appends to output if not already present — handles `self.env` vs empty dict distinction.
- Tests cover blocklist wins over passthrough, empty values ignored, local vs empty-extra env, and remote unknown case.

**Non-blocking notes**
- Terminal-tool path appends note to `output` in addition to dedicated field — intentional DX, but be aware output parsing that expects exact stdout will see extra suffix on local terminal runs; execute_code correctly does NOT append to output.

No blocking issues found.

Bartok9:
Thanks for the review — agreed on the terminal vs execute_code asymmetry: terminal appends the scrub note for local DX (models reading stdout), execute_code keeps it on the dedicated `credential_scrub_note` field only so structured output stays clean. Non-blocking note noted; happy to adjust if maintainers prefer field-only on terminal too.

PR #96048, opened 2026-08-27
fix(feishu): render markdown tables as native card table components

linyubin:
## Summary

Markdown tables in outbound Feishu messages (agent replies and cron deliveries) render inside `post`-type `md` elements with **zero column-width control**: Feishu squeezes each column to content width, so long CJK/latin cells wrap into unreadable one-or-two-character vertical strips. This PR routes table-containing chunks through Feishu's interactive-card native `table` component (JSON 2.0), which supports real width control:

- **content-aware column widths** — narrow columns get an estimated fixed px width (CJK-aware via `unicodedata.east_asian_width`, clamped to the documented [80, 600] px range), wide columns stay `"auto"` to absorb the remaining card width
- `row_height: "auto"` so rows expand instead of clipping
- cells use `lark_md` data type, keeping bold/code/links rendering inside cells

Before / after (same table):

| | |
|---|---|
| post/md path | columns ~100px, text wrapping every few characters |
| card table component | columns sized to content, rows expand to fit |

## Changes

- **`plugins/platforms/feishu/feishu_table_card.py`** (new): pipe-table → JSON 2.0 card converter. Preserves prose/table interleaving order, isolates fenced code blocks into their own markdown elements, handles alignment syntax (`:--` / `:-:` / `--:`) and escaped pipes (`\|`). Returns `None` when content has no convertible table or exceeds Feishu card limits (>5 tables/card, >50 cols/table, >100 rows/table), so callers keep the existing post path unchanged.
- **`plugins/platforms/feishu/adapter.py`**:
  - `_build_outbound_payload` gains an `allow_card_table` flag; convertible tables return an `interactive` payload
  - `send()` walks a payload downgrade ladder **interactive → post → text** when the API rejects the format (raised exception or non-success `msg`), preserving the existing post→text fallback semantics
  - `edit_message()` passes `allow_card_table=False`: Feishu's message-update API cannot change a message's `msg_type`, so edits of existing text/post messages must not become cards

Out of scope deliberately: no config knob — tables are strictly more readable in card form and all failure paths degrade to today's behavior.

## Testing

- New module unit-tested: CJK/basic tables, alignment parsing, escaped pipes, fence isolation (pipe lines inside code fences are not converted), limit returns (`>5` tables → `None`, `>50` cols → `None`), ragged rows, table-at-message-start, lark_md markup preservation
- `pytest tests/gateway/test_feishu.py` — 76 passed (no behavior change for table-free traffic: prose still goes text/post as before)
[...]

Issue #52786, opened 2026-06-26
Feishu adapter incorrectly downgrades markdown tables to plain text


frank-xd-wang:
## Summary

`plugins/platforms/feishu/adapter.py` forces any message containing a markdown table to `msg_type=text` via `_MARKDOWN_TABLE_RE` / `_build_outbound_payload`. This causes Feishu clients to render raw pipe-and-dash source instead of a table.

## Reproduction

1. Configure Feishu and start the gateway.
2. From a Feishu DM, send any message containing a markdown table, e.g.:

```
| col A | col B |
| ----- | ----- |
| 1     | 2     |
```

3. The recipient sees the source text, not a rendered table.

## Evidence that the underlying assumption is wrong

The branch is justified in the code comment as:

> Feishu post-type 'md' elements do not render markdown tables.

Empirically this is no longer true on current Feishu clients (verified 2026-06-26):

1. `lark-cli 1.0.47+ im +messages-send --markdown` sends a `post` payload with `md` elements containing a markdown table — **renders correctly**.
2. A manually constructed `post` payload matching the adapter's exact `[[{tag:"md", text:"..."}]]` shape — **renders correctly**.
3. End-to-end agent `send_message` tool via a patched adapter — **renders correctly**.

Reference message_ids (Feishu):
- `om_x100b6ce834fc50a0c228edc1f49f6dd` (lark-cli path, 08:28:19)
- `om_x100b6ce9422438a0c1078489349cab5` (manual post, 09:07:26)

## Suggested fix

Remove the `_MARKDOWN_TABLE_RE` regex and the matching branch in `_build_outbound_payload`. Let table content flow through the same `post` + `md` element path as other markdown. The net change is ~10 lines.

I have a working local patch with verified rendering; happy to send a PR if maintainers prefer that route, but wanted to surface the bug first in case the fix belongs to someone more familiar with the adapter.

## Environment

- `hermes-agent` `main` @ `7cd5eaa64`
- `feishu-platform` bundled plugin 1.0.0
- `lark-cli` 1.0.47
- Feishu clients on iOS / desktop, observed 2026-06-26

alt-glitch:
> *This was generated by AI during triage.*

[...]

PR #98358, opened 2026-08-30
feat(feishu): opt-in card tables with content-driven column widths

linyubin:
## Problem

The post/md table fix (#52786, related discussion in #96048) made tables render on Feishu clients, but the post renderer squeezes columns to equal widths: long CJK/Latin cells wrap into unreadable narrow strips and column width is not controllable.

Feishu's interactive-card `table` component (schema 2.0) supports real column-width control: `width: "auto"` (size-to-content) or a pinned px value in [80, 600].

## Change

Adds an **opt-in** `tables_as_cards` setting (`platforms.feishu.extra.tables_as_cards: true` in config.yaml, default **false** — the default post/md path is untouched) that routes table-containing outbound chunks through a new plugin-local converter:

- **`plugins/platforms/feishu/feishu_table_card.py`** — parses GFM pipe tables (escaped pipes, alignment colons, code-fence isolation) into card table components; interleaved prose is preserved as markdown elements in original order.
- **Two-pass width algorithm** — estimates each column's content width (CJK double-width aware, markdown markers discounted), then pins narrow columns to px only when at least one wide `auto` column exists to absorb the remaining card width. All-narrow tables stay fully `auto` since Feishu stretches the table to card width regardless.
- **Graceful fallback** — returns to the historical post path whenever content exceeds card component limits (>5 tables, >50 columns, >100 rows) or contains no convertible table.
- **Downgrade ladder** — the send pipeline now walks interactive → post → text when the API rejects a payload, so a rejected card can never strand a message.
- **Update-path guard** — message edits never emit cards (the im/v1 update API rejects msg_type changes).

## Testing

- New `tests/gateway/test_feishu_table_card.py` (17 tests): fallback contract, card structure/order, width bounds + pinning behavior, escaped pipes, alignment round-trip, and adapter routing (opt-in / default-off / update-path guard).
- Full related suite green: `test_feishu_table_card.py + test_feishu_table_markdown.py + test_feishu_approval_buttons.py` → 31 passed.
- E2E verified on a live gateway: real table messages sent through a Feishu DM with the flag enabled render with content-proportional column widths.

## Notes

Default stays `false`: current Feishu clients render post/md tables acceptably for most content (#52786); this is for users who need readable wide/CJK tables.

### DIFF
--- a/plugins/platforms/feishu/adapter.py
+++ b/plugins/platforms/feishu/adapter.py
@@ -66,6 +66,8 @@ from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
 from types import SimpleNamespace
+
+from plugins.platforms.feishu.feishu_table_card import build_table_card_payload
 from typing import Any, Dict, List, Literal, Optional, Sequence
 from urllib.error import HTTPError, URLError
 from urllib.parse import urlencode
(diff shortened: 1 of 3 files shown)


liyangbing:
<h3>Review</h3>

<p>The opt-in routing and downgrade ladder are clearly separated, but the parser's code-fence contract is not currently enforced at extraction time.</p>

<h3>Fenced-code parsing bug</h3>

<p><a href='https://github.com/NousResearch/hermes-agent/blob/92809eec85f122b950d62520a0f5f2b5838fe0ae/plugins/platforms/feishu/feishu_table_card.py#L105-L143'><code>_extract_table_blocks()</code> scans every line for a pipe-row header and separator</a> without tracking whether the cursor is inside a fenced code block. <a href='https://github.com/NousResearch/hermes-agent/blob/92809eec85f122b950d62520a0f5f2b5838fe0ae/plugins/platforms/feishu/feishu_table_card.py#L250-L288'><code>_prose_elements()</code> splits fences only after the table blocks have already been extracted</a>, so it cannot undo a false table match.</p>

<p>For example, a fenced snippet containing <code>| a | b |</code>, <code>| --- | --- |</code>, and a data row will currently be eligible for conversion even though it is code, not a markdown table. Please make extraction fence-aware (including language fences and a closing fence) and leave those lines in one markdown element.</p>

<h3>Regression matrix</h3>

<p>Cover a real table outside fences, a table-shaped snippet inside a fenced block, an unclosed fence, fences before and after a table, inline code containing pipes, escaped pipes, and interleaved prose. Assert that only the real table becomes a Feishu <code>table</code> element and that fenced content is byte-equivalent in the markdown fallback element.</p>

<p><strong>Scope:</strong> this review is based on Hermes PR #98358, its stated head revision, and the linked Hermes source/tests. The DeepSeek Harness handbook is an independent community project and is not Hermes documentation or a direct fix, so it is intentionally not linked here.</p>

PR #98362, opened 2026-08-30
fix(cron): honor cron.execution_retention instead of a hardcoded 1000-row cap

Navlem:
## Problem

`cron/executions.py` hardcodes the terminal-row cap:

```python
MAX_TERMINAL_EXECUTIONS = 1000
...
def _prune_unlocked(conn):
    limit = max(0, int(MAX_TERMINAL_EXECUTIONS))
```

Nothing reads `cron.execution_retention`. The key is accepted and echoed back, so the setting looks applied while pruning ignores it entirely:

```
$ hermes config set cron.execution_retention 2500
✓ Set cron.execution_retention = 2500
$ hermes config get cron.execution_retention
2500
```

…and the ledger then sits at exactly **1000** terminal rows. Verified on a live install: config `2500`, `select count(*) from executions where status in ('completed','failed')` → `1000`.

The practical cost is silent loss of operational history. At ~220 runs/day that caps the ledger at roughly **4.5 days** of coverage instead of the ~11 the operator configured, and the only symptom is that older executions quietly disappear — no warning, and `config get` actively suggests the larger value is in force.

## Change

`_terminal_execution_retention()` resolves the cap from profile-local config at prune time, so a change takes effect without a Python reinstall.

`MAX_TERMINAL_EXECUTIONS` is kept as an alias of `_DEFAULT_MAX_TERMINAL_EXECUTIONS` and doubles as a sentinel: if it differs from the default, an explicit override wins over config. That preserves the existing test hook — `tests/cron/test_execution_ledger.py` monkeypatches the constant to `3` — rather than breaking it.

`config_defaults.py` documents the key alongside the neighbouring `output_retention`, so the default is discoverable and can't silently drift from the code fallback.

Claimed/running rows are untouched, exactly as before.

## Tests

Three additions to `tests/cron/test_execution_ledger.py`:

- **the configured cap actually bounds the ledger** — exercises the real prune path against a temp `HERMES_HOME` with a cap of `7`, deliberately neither the `1000` default nor a round number, so a pass cannot be a coincidence of the old behaviour
- **an explicit `MAX_TERMINAL_EXECUTIONS` override still wins over config** — guards the existing monkeypatch hook
- **the documented default matches the module fallback** — an invariant, so the two cannot drift apart

## Verification

E2E against the real prune path (temp `HERMES_HOME`, cap 7):

```
config cap=7 -> resolved=7
wrote 20 terminal rows -> ledger holds 7
test override 3 -> resolved=3
```

Control on `main`: the same scenario fails (`AttributeError: no attribute '_DEFAULT_MAX_TERMINAL_EXECUTIONS'`), confirming the fix is load-bearing.

Mutation-verified: reintroducing the bug (`limit = max(0, int(MAX_TERMINAL_EXECUTIONS))`) makes `test_retention_reads_execution_retention_from_config` **fail** (1 failed, 21 passed); restoring it passes (22 passed). So the new guard is not tautological.

Full `tests/cron` suite via `scripts/run_tests_parallel.py` (per-file subprocess isolation, matching CI): **75 files, 949 passed, 0 failed, 1 skipped**. Re-run after rebasing onto current `main`.

## Provenance

This fix was recovered from a `hermes update` autostash that had been sitting in `git stash` since 2026-08-27 and was nearly dropped during stash cleanup. The recovered diff is unchanged apart from being rebased; the tests and the verification above are new.

### DIFF
--- a/cron/executions.py
+++ b/cron/executions.py
@@ -21,7 +21,11 @@ from hermes_time import now as _hermes_now
 # dashboard operations that temporarily enter another profile cannot leak that
 # profile's execution records into the import-time home.
 EXECUTIONS_FILE: Optional[Path] = None
-MAX_TERMINAL_EXECUTIONS = 1000
+_DEFAULT_MAX_TERMINAL_EXECUTIONS = 1000
+# Backward-compatible test override. Production reads
+# ``cron.execution_retention`` at prune time so a config change does not
+# require a Python reinstall.
+MAX_TERMINAL_EXECUTIONS = _DEFAULT_MAX_TERMINAL_EXECUTIONS
 _TERMINAL_STATES = ("completed", "failed", "unknown")
 _lock = threading.RLock()
 _PROCESS_ID = uuid.uuid4().hex
@@ -126,8 +130,25 @@ def _owner_is_live(pid: int, started_at: Optional[int]) -> bool:
     return current is not None and current == started_at
 
 
+def _terminal_execution_retention() -> int:
+    """Resolve the terminal execution-row cap from profile-local config."""
+    if MAX_TERMINAL_EXECUTIONS != _DEFAULT_MAX_TERMINAL_EXECUTIONS:
+        return max(0, int(MAX_TERMINAL_EXECUTIONS))
+    try:
+        from hermes_cli.config import load_config
+
+        cfg = load_config() or {}
+        cron_cfg = cfg.get("cron", {}) if isinstance(cfg, dict) else {}
+        value = cron_cfg.get(
+            "execution_retention", _DEFAULT_MAX_TERMINAL_EXECUTIONS
+        ) if isinstance(cron_cfg, dict) else _DEFAULT_MAX_TERMINAL_EXECUTIONS
+        return max(0, int(value))
+    except (TypeError, ValueError):
+        return _DEFAULT_MAX_TERMINAL_EXECUTIONS
+
+
 def _prune_unlocked(conn: sqlite3.Connection) -> None:
-    limit = max(0, int(MAX_TERMINAL_EXECUTIONS))
+    limit = _terminal_execution_retention()
     conn.execute(
         """DELETE FROM executions WHERE id IN (
              SELECT id FROM executions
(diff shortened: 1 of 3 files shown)


liyangbing:
<h3>Review</h3><p>This fixes the most important contract violation: <code>config set/get</code> can claim a retention value while the prune path still silently caps terminal history at 1000 rows. The temp-profile and mutation verification are strong evidence that the configured cap is actually consulted.</p><p>Please keep the retention boundary explicit:</p><p>1. Resolve the value from the same profile-local configuration source used by <code>config get</code>, at prune time, with a documented precedence rule for the legacy test override.<br>2. Validate malformed, negative, zero, fractional, and extremely large values without turning a configuration error into silent data loss or an unbounded query.<br>3. Make the effective cap observable in diagnostics/health output, including the source of the value (config, default, or explicit test/legacy override).<br>4. Preserve claimed/running rows and define whether a retention decrease is applied immediately, including crash/restart behavior during pruning.<br>5. Ensure concurrent pruners cannot race into an inconsistent count or delete rows newly transitioned to a non-terminal state.</p><h3>Regression matrix</h3><p>Please cover:<br>• configured cap 7, default cap, and explicit legacy override<br>• config changed between two prune calls without reinstall/restart<br>• missing, malformed, negative, zero, and very large retention values<br>• completed/failed rows mixed with claimed/running rows<br>• concurrent prune and status transition, crash/restart during prune, and idempotent repeated prune<br>• profile isolation so one <code>HERMES_HOME</code> cannot change another ledger’s retention</p><p>Direct evidence: <a href="https://github.com/NousResearch/hermes-agent/pull/98362">PR #98362</a>, which changes <code>cron/executions.py</code>, documents the key in <code>hermes_cli/config_defaults.py</code>, and adds ledger tests.</p>

Navlem:
Closing in favour of #97818, which now carries the combined fix.

Both PRs addressed the same bug, and neither was complete on its own:

- This PR's `MAX_TERMINAL_EXECUTIONS != _DEFAULT_MAX_TERMINAL_EXECUTIONS` sentinel silently ignores an override set to exactly `1000` — a test or embedder pinning the cap to the default value falls through to config instead. #97818 uses a `None` sentinel, which does not have that hole. There is now a regression test (`test_retention_override_equal_to_default_is_honoured`) that fails against the `!= _DEFAULT` form.
- Conversely, #97818 originally omitted this PR's `hermes_cli/config_defaults.py` entry, without which `hermes config set cron.execution_retention` does not recognise the key. That registration has been carried over, along with a test asserting it.

#97818 now contains both halves plus the docs change. No work from this PR is lost.

PR #98365, opened 2026-08-30
fix(launcher): self-relaunch under the venv interpreter when invoked directly

DENSENSE8:
## Problem

`hermes` (the repo-root CLI launcher) has shebang `#!/usr/bin/env python3`, which resolves to whatever `python3` is first on PATH — not Hermes' venv. The installed `~/.local/bin/hermes` wrapper always execs this file via `venv/bin/python <repo>/hermes`, so the shebang mismatch is invisible on a normal terminal launch.

But `resolve_hermes_bin()` (`hermes_cli/relaunch.py`) prefers `sys.argv[0]`, and a regenerated `.desktop` entry (`linux_desktop_entry.py`) or a relaunch via `build_relaunch_argv()` can point straight at this repo-root script and exec it directly — under the system python, which lacks Hermes' deps. It dies on the first third-party import (#90292), silently, since `Terminal=false` swallows the traceback: launching Hermes from a menu/launcher does nothing, with no visible error.

## Fix

When this script isn't already running under `<repo>/venv/bin/python`, re-exec itself under that interpreter, resolved relative to the script's own real location at runtime (`Path(__file__).resolve().parent / "venv" / "bin" / "python"`) — so it's correct for any checkout path or user, not tied to one machine.

## Verified

Reproduced the failure and confirmed the fix live: running `python3 hermes --version` under the bare system interpreter (no venv deps on its import path) now correctly re-execs into the venv and prints the normal version banner, instead of `ModuleNotFoundError`.

Companion to # which fixes a related but distinct symptom in the same area (a stale resolved interpreter path baked into the `.desktop` entry after a venv-generation rotation).

### DIFF
--- a/hermes
+++ b/hermes
@@ -6,6 +6,32 @@ This wrapper should behave like the installed `hermes` command, including
 subcommands such as `gateway`, `cron`, and `doctor`.
 """
 
+import os
+import sys
+from pathlib import Path
+
+# Re-exec under the venv interpreter when invoked directly (not via the
+# installer's `venv/bin/python <repo>/hermes` wrapper). `#!/usr/bin/env
+# python3` — required for this script to be portable across installs — picks
+# up whatever `python3` is first on PATH, which usually lacks Hermes' venv
+# deps. That mismatch is invisible on a normal terminal launch (the installer
+# wrapper always runs this under the venv), but a regenerated `.desktop`
+# entry can point straight at this file (see linux_desktop_entry.py's
+# `resolve_hermes_bin()` / `sys.argv[0]` preference), which then dies on the
+# first third-party import (#90292) — silently, since `Terminal=false`
+# swallows the traceback.
+#
+# Resolved at runtime relative to this file's own real location, so it works
+# for any checkout path / any user, not just the one that first hit the bug.
+_venv_python = Path(__file__).resolve().parent / "venv" / "bin" / "python"
+if _venv_python.is_file():
+    try:
+        _running_under_venv = Path(sys.executable).resolve() == _venv_python.resolve()
+    except OSError:
+        _running_under_venv = False
+    if not _running_under_venv:
+        os.execv(str(_venv_python), [str(_venv_python), str(Path(__file__).resolve())] + sys.argv[1:])
+
 if __name__ == "__main__":
     from hermes_cli.main import main
     main()


liyangbing:
<h3>Review</h3><p>The fix addresses the direct-launch identity gap, but the relaunch contract needs a few explicit boundaries before it is safe across managed and unmanaged installs. The important invariant is that the launcher must execute the same environment whose package and profile it is meant to operate, while avoiding a relaunch loop or silently selecting a different checkout.</p><p>Please make these cases explicit in the implementation and tests:</p><p>1. Detect the active interpreter using normalized real paths, including Windows case/separator differences and symlinked <code>sys.executable</code>; do not compare raw strings.<br>2. Resolve the venv relative to the intended repository root, then verify that the candidate interpreter exists and is executable before re-execing.<br>3. Preserve all original argv, environment/profile selection, and exit status; a failed re-exec should surface a diagnostic rather than silently exiting from a <code>Terminal=false</code> launcher.<br>4. Guard against an absent venv, a broken symlink, a non-venv interpreter, and repeated self-relaunch when the candidate resolves back to the current interpreter.<br>5. Keep the system-Python fallback behavior explicit for source checkouts without a managed venv, rather than making the fallback depend on an incidental PATH entry.</p><h3>Regression matrix</h3><p>Please cover:<br>• direct <code>python3 hermes --version</code> outside the venv and normal installed-wrapper launch<br>• current interpreter already inside the venv: no second exec<br>• symlinked venv interpreter and resolved system interpreter<br>• repository path containing spaces, non-ASCII characters, or Windows separators<br>• missing/broken venv interpreter and preserved non-zero exit evidence<br>• arbitrary CLI arguments, profile variables, and signal/exit-code propagation across the re-exec<br>• desktop launch through the generated <code>Exec=</code> command, including the companion stable-path update rotation case</p><p>Direct evidence: <a href="https://github.com/NousResearch/hermes-agent/pull/98365">PR #98365</a>, which changes <code>hermes</code> and documents the <code>resolve_hermes_bin()</code> / <code>build_relaunch_argv()</code> launch paths.</p>

estebanmunchjones2019:
Additional reproduction from a Linux source checkout:

- Host: RHEL 9, `/usr/bin/python3` = Python 3.9.21
- Checkout declares `requires-python = ">=3.11,<3.14"`
- Running the repo-root launcher directly (`python3 hermes sessions browse`) imports `hermes_cli.main` with the host interpreter and fails during annotation evaluation:

```text
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

The installed shell shim works because it explicitly invokes the managed Hermes venv Python. This confirms the issue is broader than the macOS desktop path: direct/source-tree launcher execution can bypass the supported interpreter. I found PR #98365, which appears to address the same launcher-level root cause by self-relaunching under the checkout's venv interpreter, so I am not opening a duplicate PR.

PR #98367, opened 2026-08-30
fix(desktop-entry): use the stable venv/bin/python symlink, not a resolved generation path

DENSENSE8:
## Problem

`resolve_exec_command()` fully resolves `sys.executable` when it needs to prefix the `.desktop` entry's `Exec=` with an interpreter. On installs that manage Python via rotating "generation" directories (`venv/bin/python` as a symlink that gets atomically repointed on update), `.resolve()` follows that symlink all the way down and bakes the literal, generation-pinned path into the installed `~/.local/share/applications/hermes.desktop`.

`hermes update` rotates to a new generation and prunes the old one, so the already-installed `.desktop` entry then points at a deleted binary. A launcher menu (Walker/wofi/etc.) selecting "Hermes" execs nothing — silently, since `Terminal=false` swallows the error. The entry only self-heals the next time `hermes desktop` happens to be run by hand (which regenerates it) — chicken-and-egg if the launcher menu is the only way Hermes gets opened.

## Fix

Prefer the stable, unresolved `<repo>/venv/bin/python` path — the one `hermes update` atomically repoints rather than deletes — over a fully resolved `sys.executable`. Falls back to the previous behavior when that venv path doesn't exist (e.g. no managed venv at all).

Companion to #98365, which fixes a related but distinct symptom in the same area (the repo-root `hermes` script itself lacking a venv-aware shebang).

### DIFF
--- a/hermes_cli/linux_desktop_entry.py
+++ b/hermes_cli/linux_desktop_entry.py
@@ -78,7 +78,22 @@ def resolve_exec_command() -> str:
             # third-party import (#90292) — silently, since Terminal=false.
             # sys.executable is the interpreter actually running Hermes (the
             # venv one), so prefix it explicitly.
-            argv = [str(Path(sys.executable).resolve()), str(resolved), "desktop"]
+            #
+            # LOCAL PATCH (see ~/.hermes/local-patches/): don't fully
+            # .resolve() sys.executable — that follows the venv/bin symlink
+            # all the way down into the current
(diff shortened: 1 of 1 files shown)


alt-glitch:

Duplicate of #96777: both preserve the unresolved venv interpreter in the generated desktop Exec= path. #96777 is earlier and covers the same update-rotation failure more comprehensively.

PR #3854, opened 2026-03-30
fix(gateway): pass full transcript to compressor instead of filtered messages

Git-on-my-level:
## Problem

The gateway `/compress` command (manual) and session hygiene auto-compress both filter messages to only `user`/`assistant` roles before passing to the context compressor. This causes:

1. **No compression for shorter conversations** — the compressor's early-return guard (`n_messages <= protect_first_n + protect_last_n + 1`) triggers because the filtered count is too low
2. **Tool results never pruned** — tool outputs (often the biggest token consumers) are already stripped, so `_prune_old_tool_results()` has nothing to work on
3. **Misleading stats** — before/after counts and tokens are computed on the filtered list, showing identical numbers (e.g. `🗜 Compressed: 6 → 6 messages`)

## Fix

Pass `list(history)` (full transcript with all roles preserved) to `_compress_context` in both locations:

- `_handle_compress_command` (manual `/compress`)
- Session hygiene auto-compress in `_handle_message_with_agent`

The compressor already handles tool messages correctly via `_prune_old_tool_results()` and `_sanitize_tool_pairs()`. The gateway filtering was defeating those mechanisms.

## Testing

Added two new tests:
- `test_session_hygiene_passes_full_history_to_compressor` — verifies auto-compress passes full history including tool messages
- `test_manual_compress_passes_full_history_to_compressor` — verifies manual `/compress` passes full history including tool messages

All 21 session hygiene tests pass.

PR #98368, opened 2026-08-30
fix(gateway): /compress --preview undercounts by dropping tool results, system prompt and tool schemas

### LINKED ISSUE #98360: [Bug]: gateway /compress --preview undercounts by ~100x vs CLI — same shared reporter, different inputs (system prompt, tools and tool results dropped)
## Bug Description

`summarize_compress_preview` exists specifically so the two `/compress --preview`
surfaces agree. Its docstring (`hermes_cli/partial_compress.py:152-156`) says:

> Shared by the CLI (``cli.py::_manual_compress``) and the gateway
> (``gateway/slash_commands.py::_handle_compress_command``) **so both surfaces report
> the same numbers the real run would use.**

The two call sites pass different inputs, so the numbers cannot agree.

**CLI** (`cli.py`, preview branch) passes the full history plus the system prompt and
the tool schemas:

```python
_sys_prompt = getattr(self.agent, "_cached_system_prompt", "") or ""
_tools = getattr(self.agent, "tools", None) or None
approx_tokens = estimate_request_tokens_rough(
    self.conversation_history,
    system_prompt=_sys_prompt,
    tools=_tools,
)
report = summarize_compress_preview(self.conversation_history, ...)
```

**Gateway** (`gateway/slash_commands.py`, preview branch) filters by role and passes
neither:

```python
_pv_msgs = [
    {"role": m.get("role"), "content": m.get("content")}
    for m in history
    if m.get("role") in {"user", "assistant"} and m.get("content")
]
approx_tokens = estimate_request_tokens_rough(_pv_msgs)
report = summarize_compress_preview(_pv_msgs, partial, keep_last, focus_topic, approx_tokens)
```

Three payload buckets drop out on the gateway path:

- the **system prompt** and the **tool schemas** — the blind spot
  `estimate_request_tokens_rough` warns about in its own docstring
  (*"With 50+ tools enabled, schemas alone can add 20-30K tokens"*);
- the **tool results** (`role: "tool"`), excluded by the role filter — even though
  compaction does rewrite them.

Because the filtered list is also what `summarize_compress_preview` receives, the
message counts diverge too: `Would compress N of M message(s)` counts only
user+assistant rows on the gateway and every row on the CLI.

The user-visible string is `~{approx_tokens:,} tokens currently in context`, so on the
gateway a number that is not the context is labelled as the context.

### Deterministic, offline reproduction

Run inside the released image (`nousresearch/hermes-agent:v2026.8.27`); no gateway, no
network, no session required:

```python
from agent.model_metadata import estimate_request_tokens_rough

history = [
    {"role": "user", "content": "read the config and summarise it"},
    {"role": "assistant", "content": "", "tool_calls": [{"id": "c1", "function": {"name": "terminal", "arguments": "{}"}}]},
    {"role": "tool", "tool_call_id": "c1", "content": "X" * 4000},
    {"role": "assistant", "content": "Here is the summary."},
    {"role": "user", "content": "now check the logs"},
    {"role": "assistant", "content": "", "tool_calls": [{"id": "c2", "function": {"name": "terminal", "arguments": "{}"}}]},
    {"role": "tool", "tool_call_id": "c2", "content": "Y" * 4000},
    {"role": "assistant", "content": "The logs are clean."},
]
system_prompt = "S" * 8000
tools = [{"type": "function", "function": {"name": "t%d" % i, "description": "D" * 300,
          "parameters": {"type": "object", "properties": {}}}} for i in range(20)]

# what the GATEWAY passes
pv = [{"role": m.get("role"), "content": m.get("content")}
      for m in history
      if m.get("role") in {"user", "assistant"} and m.get("content")]
print("gateway:", len(pv), "messages ->", estimate_request_tokens_rough(pv), "tokens")

# what the CLI passes
print("cli    :", len(history), "messages ->",
      estimate_request_tokens_rough(history, system_prompt=system_prompt, tools=tools), "tokens")
```

Output:

```
gateway: 4 messages -> 57 tokens
cli    : 8 messages -> 5823 tokens
```

Same state, same reporter, **102x apart** — and 4 messages vs 8.

### On a live session

On a real gateway session (Hermes `v2026.8.27`, `gpt-5.6-terra` on the Codex OAuth
route), re-running the gateway's own filter over the persisted transcript reproduces the
reported figures exactly:

```
roles in transcript:        {'user': 28, 'assistant': 53, 'tool': 31, 'session_meta': 1}

user/assistant only:        55 messages -> 1890 tokens      <- what /compress --preview printed
all roles with content:     86 messages -> 7000 tokens
actual context (compressor last_prompt_tokens): 36730 tokens
```

`55` and `1890` are byte-identical to what the command printed:

```
🗜️ Would compress 55 of 55 message(s) (~1,890 tokens currently in context).
```

So on that session the preview reported **5%** of the context it named, and the largest
missing slice is the system prompt plus tool schemas.

## Expected Behavior

Either the gateway preview reports the same figure the CLI reports for the same session
state, or the label stops calling it the context.

The contract that makes this a bug rather than a preference is the shared reporter's own
docstring: it exists *"so both surfaces report the same numbers the real run would use."*

## Version

  release at the time of writing.
- Still present on `main` at `2635035` (2026-08-30 03:57 UTC): the gateway filter is
  unchanged at `gateway/slash_commands.py:4496-4508`, the CLI call at `cli.py:13843-13858`,
  and the docstring at `hermes_cli/partial_compress.py:152-156`.

Run `/compress --preview` and `/context` in the same gateway session, back to back. If
the preview's `tokens currently in context` and `/context`'s `In use:` differ by an order
of magnitude, it is back.

## Not a duplicate of #92570

#92570 is a different failure on the same command — the Desktop/TUI path treating
`--preview` as a focus string and performing a real compression. It quotes this report
format as the *expected* output, including `(~X tokens currently in context)`, so it does
not cover the figure itself.

chelsealong:
Fixes #98360.

## Problem

`summarize_compress_preview`'s own docstring (`hermes_cli/partial_compress.py:152-156`)
says it exists so the CLI's `/compress --preview` and the gateway's agree on the
reported numbers. They didn't, because the two call sites fed it different inputs:

- **CLI** (`cli.py::_manual_compress`) passes the full history plus the cached system
  prompt and the agent's tool schemas to `estimate_request_tokens_rough`.
- **Gateway** (`gateway/slash_commands.py::_handle_compress_command_inner`, preview
  branch) filtered the transcript to `user`/`assistant` messages with content and
  passed neither `system_prompt` nor `tools`.

That drops three payload buckets on the gateway path: the system prompt, the tool
schemas (which the function's own docstring warns can be 20-30K tokens with 50+ tools
enabled), and every `tool`-role message (tool results), even though the real
(non-preview) compression branch a few lines below already special-cases keeping tool
messages for exactly this reason (`#3854`). On a real session the issue reports the
preview printed `~1,890 tokens currently in context` while the actual context
(`last_prompt_tokens`) was `36,730` — an order of magnitude off, and the label calls it
"the context" while missing most of it.

## Fix

`gateway/slash_commands.py`, preview branch: mirror the real compression path a few
lines below it instead of diverging from it.

- Keep `tool`-role messages in the filtered transcript (matches the non-preview `msgs`
  filter right below, and the `#3854` rationale in its comment).
- Restore the session's persisted system prompt from `self._session_db.get_session()`
  — the same technique `gateway/run.py::_seed_hygiene_system_prompt` already uses, so
  no new agent has to be built just to read it back.
- Resolve the tool schemas a live turn for this session would actually send, via the
  same `_resolve_enabled_toolsets_for_source()` + `model_tools.get_tool_definitions()`
  pair the real gateway turn path already calls (`gateway/run.py:29319-29328`), instead
  of spinning up a full `AIAgent` (which the preview branch deliberately avoids — "no
  agent, no writes").

All three lookups are wrapped so a failure degrades to the old (partial) numbers
instead of breaking `--preview`.

## Testing

Added `tests/gateway/test_compress_preview.py::test_preview_feeds_tool_messages_system_prompt_and_tools_to_estimator`,
which asserts the shared estimator is actually called with tool-result messages, a
non-empty system prompt, and the resolved tool schemas.

Confirmed it fails without the fix:

```
$ git checkout HEAD~1 -- gateway/slash_commands.py   # pre-fix version
$ python -m pytest tests/gateway/test_compress_preview.py -q
..F
AssertionError: tool-result messages dropped from the preview estimate: ['user', 'assistant', 'user', 'assistant']
assert 'tool' in ['user', 'assistant', 'user', 'assistant']
1 failed, 2 passed in 1.47s

$ git checkout HEAD -- gateway/slash_commands.py     # fix restored
$ python -m pytest tests/gateway/test_compress_preview.py -q
...
3 passed in 2.64s
```

Full suite for the touched command, all green:

```
$ python -m pytest tests/gateway/test_compress_preview.py tests/gateway/test_compress_command.py \
    tests/gateway/test_compress_focus.py tests/gateway/test_compress_plugin_engine.py -q
13 passed in 3.64s
```

`ruff check gateway/slash_commands.py tests/gateway/test_compress_preview.py` — all
checks passed.

## Disclosure

Investigated, written, and tested with AI assistance (Claude).

### DIFF
--- a/gateway/slash_commands.py
+++ b/gateway/slash_commands.py
@@ -4491,14 +4491,62 @@ class GatewaySlashCommandsMixin:
                 return _agg_note
 
         if _preview:
-            # Report what WOULD be compressed — no agent, no writes.
+            # Report what WOULD be compressed — no agent, no writes. Mirror
+            # the real compression path below: keep tool-result messages
+            # (compaction rewrites them too) and fold in the system prompt
+            # + tool schemas, so this matches cli.py's preview branch and
+            # the two surfaces report the same figure for the same session
+            # state (#98360) instead of the gateway silently dropping the
+            # three biggest payload buckets.
             from agent.model_metadata import estimate_request_tokens_rough
+            from agent.skill_utils import parse_config_string_list
+            from gateway.run import _load_gateway_config, _platform_config_key
+            from model_tools import get_tool_definitions
+
             _pv_msgs = [
-                {"role": m.get("role"), "content": m.get("content")}
-                for m in history
-                if m.get("role") in {"user", "assistant"} and m.get("content")
+                m for m in history
+                if m.get("role") in {"user", "assistant", "tool"}
             ]
-            approx_tokens = estimate_request_tokens_rough(_pv_msgs)
+
+            _sys_prompt = ""
+            _get_session = getattr(self._session_db, "get_session", None)
+            if callable(_get_session):
+                try:
+                    _pv_session_row = await _get_session(session_entry.session_id)
+                except Exception:
+                    _pv_session_row = None
+                if isinstance(_pv_session_row, dict):
+                    _raw_prompt = _pv_session_row.get("system_prompt")
+                    if isinstance(_raw_prompt, str) and _raw_prompt.strip():
+                        _sys_prompt = _raw_prompt
+
+            _pv_tools = None
+            if source.platform:
+                try:
+                    _pv_cfg = _load_gateway_config()
+                    _pv_platform_key = _platform_config_key(source.platform)
+                    _pv_enabled = self._resolve_enabled_toolsets_for_source(
+                        _pv_cfg, source, _pv_platform_key
+                    )
+                    _pv_disabled = parse_config_string_list(
+                        (_pv_cfg.get("agent") or {}).get("disabled_toolsets")
+                    ) or None
+                    _pv_tools = get_tool_definitions(
+                        enabled_toolsets=_pv_enabled,
+                        disabled_toolsets=_pv_disabled,
+                        quiet_mode=True,
+                    ) or None
+                except Exception:
+                    logger.warning(
+                        "Compress preview could not resolve tool schemas for "
+                        "session %s; estimate will exclude tool schemas.",
+                        session_entry.session_id,
+                        exc_info=True,
+                    )
+
+            approx_tokens = estimate_request_tokens_rough(
+                _pv_msgs, system_prompt=_sys_prompt, tools=_pv_tools
+            )
             report = summarize_compress_preview(
                 _pv_msgs, partial, keep_last, focus_topic, approx_tokens
             )
--- a/gateway/slash_commands.py
+++ b/gateway/slash_commands.py
@@ -4508,6 +4508,11 @@ class GatewaySlashCommandsMixin:
                 if m.get("role") in {"user", "assistant", "tool"}
             ]
 
+            # Track which payload buckets we couldn't resolve, so a lookup
+            # failure degrades to a labeled lower bound instead of silently
+            # posing as the full context (reviewer feedback on #98368).
+            _omitted_buckets = []
+
             _sys_prompt = ""
             _get_session = getattr(self._session_db, "get_session", None)
             if callable(_get_session):
@@ -4515,10 +4520,13 @@ class GatewaySlashCommandsMixin:
                     _pv_session_row = await _get_session(session_entry.session_id)
                 except Exception:
                     _pv_session_row = None
+                    _omitted_buckets.append("system prompt")
                 if isinstance(_pv_session_row, dict):
                     _raw_prompt = _pv_session_row.get("system_prompt")
                     if isinstance(_raw_prompt, str) and _raw_prompt.strip():
                         _sys_prompt = _raw_prompt
+            else:
+                _omitted_buckets.append("system prompt")
 
             _pv_tools = None
             if source.platform:
@@ -4543,6 +4551,9 @@ class GatewaySlashCommandsMixin:
                         session_entry.session_id,
                         exc_info=True,
                     )
+                    _omitted_buckets.append("tool schemas")
+            else:
+                _omitted_buckets.append("tool schemas")
 
             approx_tokens = estimate_request_tokens_rough(
                 _pv_msgs, system_prompt=_sys_prompt, tools=_pv_tools
@@ -4551,6 +4562,11 @@ class GatewaySlashCommandsMixin:
                 _pv_msgs, partial, keep_last, focus_topic, approx_tokens
             )
             lines = [f"🗜️ {line}" for line in report["lines"]]
+            if _omitted_buckets:
+                lines.append(
+                    "⚠️ Lower bound — could not resolve: "
+                    + ", ".join(_omitted_buckets) + "."
+                )
             if _aggressive:
                 lines.append(_agg_note)
             return "\n".join(lines)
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related to #98360 and narrower PR #98370. This patch covers tool rows plus the system-prompt and tool-schema inputs that #98370 intentionally leaves out.

liyangbing:
<h3>Review</h3><p>This closes the major accounting gap identified in #98360: tool results, the persisted system prompt, and live tool schemas now reach the shared estimator. The remaining correctness boundary is snapshot consistency. A preview must say what context it measured, especially because the session, enabled toolsets, and provider configuration can change while the command is running.</p><p>Please make these invariants explicit:</p><p>1. Build one coherent input snapshot for transcript, persisted system prompt, enabled toolsets, tool definitions, model, and profile; do not mix values from different revisions.<br>2. Keep preview read-only: schema resolution, session reads, and fallback handling must not create an agent, write Session state, consume a provider request, or mutate tool/plugin lifecycle.<br>3. Mark partial estimates when any system/schema lookup fails, and expose which buckets were omitted; a lower-bound number must not be labeled as the full context.<br>4. Match the real compression serialization and filtering rules, including tool-call/result identity, ordering, attachments, retries, and empty-content turns.<br>5. Define behavior when configuration changes between preview and compression: report the measured revision and avoid implying that the preview guarantees the later commit threshold.</p><h3>Regression matrix</h3><p>Please cover:<br>• full snapshot with system prompt, tool schemas, tool results, empty tool calls, and mixed ordering<br>• missing/malformed system prompt, unavailable tool registry, schema-resolution failure, and partial lower-bound output<br>• concurrent Session/config/tool changes during preview, with revision or snapshot identity in the result<br>• no-agent/no-write/no-network side-effect assertions<br>• tool enablement, profile isolation, model/tokenizer differences, and large schema payloads<br>• preview versus real compression input parity and threshold decisions, including retry/rollback and attachments</p><p>Direct evidence: <a href="https://github.com/NousResearch/hermes-agent/pull/98368">PR #98368</a>, which changes <code>gateway/slash_commands.py</code> and adds the estimator-input regression in <code>tests/gateway/test_compress_preview.py</code>.</p>

chelsealong:
Addressed invariant #3 (partial estimates must not pose as the full context): `slash_commands.py`'s preview branch now tracks which of the system-prompt / tool-schema lookups failed and appends a `⚠️ Lower bound — could not resolve: …` line naming them, instead of silently falling back to the old partial numbers. Covered by a new regression test (`test_preview_labels_lower_bound_when_tool_schema_lookup_fails`); full suite for the touched files still green (14 passed) and `ruff check` clean.

The remaining invariants (#1 single coherent input snapshot, #4 full serialization parity with the real compression path incl. retries/attachments, #5 revision reporting for concurrent config changes) and the full regression matrix are a much larger scope than this bug-fix PR — happy to pick those up as follow-up work if a maintainer wants them tracked separately.

PR #98370, opened 2026-08-30
fix(gateway): count tool results in /compress --preview

bsdnn:
## What does this PR do?

The gateway `/compress --preview` branch filtered the transcript to
user+assistant messages with non-empty content before handing it to the shared
`summarize_compress_preview` reporter — dropping tool results and empty-content
`tool_call` turns. Both the token estimate and the "Would compress N of M
message(s)" count undercounted relative to the CLI and the real compression
run, which pass the full user/assistant/tool transcript (#98360).

Fix: pass the same `{user, assistant, tool}` rows the non-preview branch and the
real run use, so tool results and tool_call turns are counted.

**Out of scope (deliberately):** the system prompt and tool schemas — the
*largest* missing slice — are still not counted in the preview, because the
preview path builds no agent to obtain them. That's a separate, larger change
(or a relabel of the shared reporter) and is left for maintainer decision; this
PR closes the clean, verifiable role-filter half.

## Related Issue

Addresses the role-filter half of #98360 (kept as `Addresses`, not `Fixes`).

## Type of Change

## Changes Made

- `gateway/slash_commands.py` — preview branch counts `{user, assistant, tool}`
  rows (tool results + tool_call turns included), matching the non-preview
  branch and the CLI.
- `tests/gateway/test_compress_preview.py` — a tool-result transcript is counted
  in full (`4 of 4`, not `2 of 2`).

## How to Test

    pytest tests/gateway/test_compress_preview.py -q

3 pass. A transcript with a tool result + an empty-content tool_call turn now
previews all four compressible rows instead of the two user/assistant rows.

## Checklist

### DIFF
--- a/gateway/slash_commands.py
+++ b/gateway/slash_commands.py
@@ -4493,10 +4493,17 @@ class GatewaySlashCommandsMixin:
         if _preview:
             # Report what WOULD be compressed — no agent, no writes.
             from agent.model_metadata import estimate_request_tokens_rough
+            # Count the same rows the real run compresses: user/assistant/tool
+            # with their full payloads (tool results and tool_call turns
+            # included), matching the non-preview branch below and the CLI.
+            # The previous user+assistant, content-only filter dropped tool
+            # results and empty-content tool_call turns, so both the token
+            # estimate and the "N of M" message count undercounted (#98360).
+            # (The system prompt + tool schemas remain out of scope here — the
+            # preview path deliberately builds no agent to obtain them.)
             _pv_msgs = [
-                {"role": m.get("role"), "content": m.get("content")}
-                for m in history
-                if m.get("role") in {"user", "assistant"} and m.get("content")
+                m for m in history
+                if m.get("role") in {"user", "assistant", "tool"}
             ]
             approx_tokens = estimate_request_tokens_rough(_pv_msgs)
             report = summarize_compress_preview(
--- a/tests/gateway/test_compress_preview.py
+++ b/tests/gateway/test_compress_preview.py
@@ -85,3 +85,30 @@ async def test_aggressive_dry_run_shows_preview_plus_note():
     runner.session_store.rewrite_transcript.assert_not_called()
 
 
+def _make_tool_history() -> list[dict]:
+    return [
+        {"role": "user", "content": "read the config"},
+        {"role": "assistant", "content": "",
+         "tool_calls": [{"id": "c1", "function": {"name": "terminal", "arguments": "{}"}}]},
+        {"role": "tool", "tool_call_id": "c1", "content": "X" * 2000},
+        {"role": "assistant", "content": "done."},
+    ]
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


alt-glitch:

Related to #98360 and broader PR #98368. This patch fixes tool-row accounting only; #98368 additionally restores the system prompt and tool-schema inputs.

liyangbing:
<h3>Review</h3><p>The role-filter fix is a useful, verifiable increment, and keeping the issue as “Addresses” is accurate. The remaining contract should be visible to users: a preview that counts user, assistant, and tool rows but omits the system prompt and tool schemas must not present its estimate as equivalent to the real compression run.</p><p>Please make the approximation boundary explicit:</p><p>1. Share one canonical transcript/input builder between preview, the real compression path, and the CLI, or record exactly which inputs each path includes.<br>2. Label counts and token estimates with their scope, including whether system instructions, tool schemas, attachments, and metadata are excluded.<br>3. Keep empty-content tool-call turns and tool results paired by identity/order; a row count alone must not imply valid compression input.<br>4. Ensure preview uses the same model/tokenizer and serialization rules as the eventual compression path when available, with a clear fallback when no agent is constructed.<br>5. Preserve the fail-safe behavior when preview cannot obtain system/schema inputs: report unknown or lower-bound estimates rather than claiming a complete forecast.</p><h3>Regression matrix</h3><p>Please cover:<br>• user/assistant-only, tool-result, empty-content tool-call, and mixed interleaved transcripts<br>• system prompt and tool schemas present, absent, oversized, malformed, and dynamically changed<br>• preview versus real-run row selection, serialized bytes, token estimate, and compression threshold decision<br>• duplicate/missing tool IDs and out-of-order tool results<br>• attachment/media metadata, retries, and partial transcript recovery<br>• UI copy and machine-readable output clearly distinguishing a lower-bound preview from a full estimate</p><p>Direct evidence: <a href="https://github.com/NousResearch/hermes-agent/pull/98370">PR #98370</a>, which changes <code>gateway/slash_commands.py</code> and adds <code>tests/gateway/test_compress_preview.py</code>; the PR explicitly leaves system prompt and tool schemas for a separate change.</p>

kvnloo:
Checked out exact head d5fd1e2fc8bd09ef129bad5961cbf020ea33eb04 on Linux 7.2.2-1-cachyos x86_64.

`python3 -m pytest tests/gateway/test_compress_preview.py -q` → 3 passed in 2.52s
(log sha256 `5d4440d3a0d1423ac6421a9bea6a2c0a1b02a1d47e0822d5dcf2197caf3e6dc1`).

The new test asserts `4 of 4` and not `2 of 2` on a user / empty-content tool_call / 2000-char tool result / assistant transcript, and `rewrite_transcript` is not called.

`estimate_request_tokens_rough` is invoked without `system_prompt`/`tools`, which matches the PR’s stated out-of-scope note. Passing the full history dicts, including `tool_calls`, is the right shape for `estimate_messages_tokens_rough`.

I did not run a live gateway `/compress --preview` against a real session; only the unit file above.

Issue #88102, opened 2026-08-17
TUI: auto-compaction status never appears in transcript (kind mismatch 'compacting'/'compacted' vs 'compressing'); aborted compactions emit nothing


mayqhw:
## Summary

Automatic context compaction gives the TUI user no persistent signal that compaction is running or done. Compaction status lines only flash in the status bar for ~4 seconds (`restoreStatusAfter(4000)`) and never appear in the transcript. During a long compaction (often 1–5 minutes) there is no way to tell whether compaction is still running, completed, or silently aborted.

## Repro

1. Let a session grow until preflight/hygiene triggers auto-compaction (or a `context too large` path).
2. Observe the status bar: a `🗜️ Compacting context — …` line flashes briefly (~4 s) and disappears.
3. Contrast with manual `/compress`, which injects `⠋ compressing N messages …` into the transcript.

## Root cause

Kind mismatch between the two compaction paths in the TUI event handler (`ui-tui/src/app/createGatewayEventHandler.ts`):

- **Manual** `/compress` emits `status.update` with `kind: 'compressing'` (`tui_gateway/methods_session.py`) — the TUI matches this and injects it into the transcript via `sys()`.
- **Automatic** compaction emits:
  - `kind: 'compacting'` — `tui_gateway/server.py` `_status_update` re-tags agent `lifecycle` statuses containing the `Compacting context` marker (introduced in 715b69172 for the desktop "Summarizing…" indicator), and
  - `kind: 'compacted'` — `agent/conversation_compression.py` `_emit_compaction_done`.

Neither matches `p.kind === 'compressing'`, so auto-compaction statuses fall through to the generic branch (status-bar blip + activity note, then restore after 4 s). Additionally, an **aborted** compaction emits no status at all — the user sees nothing.

## Related UX gap

The manual `/compress` transcript line starts with a static `⠋` (braille spinner frame) that never animates — the transcript is a text snapshot; animation only runs in the status bar (`appChrome.tsx` spinner tick). So even the visible path does not convey progress or completion.

## Possible fixes (for maintainer judgment)

- Frontend: include `compacting`/`compacted` kinds in the transcript path (minimal; implemented on branch `fix/tui-compaction-status-visible` in https://github.com/mayqhw/hermes-agent, PR https://github.com/NousResearch/hermes-agent/pull/88098 — closed at author's request).
[...]

PR #70365, opened 2026-07-23
fix(tui): keep automatic compaction status visible for the whole operation (#70338)

Drexuxux:
## What
During automatic preflight/mid-turn context compaction the Ink TUI briefly shows a useful `🗜️ Compacting context — summarizing…` status, then replaces it with a generic status after ~4s — even though automatic compaction routinely runs for tens of seconds to minutes. A healthy long-running compaction therefore looks frozen, with no persistent indication that Hermes is compacting rather than stalled. Fixes #70338.

## Fix
`createGatewayEventHandler` only pinned `kind='compressing'` (manual `/compress`); an automatic `kind='compacting'` event fell through to the generic activity path and got `restoreStatusAfter(4000)`, reverting the label after four seconds.

Treat `kind='compacting'` as a first-class long-running status: keep its label pinned until the next status/content event on resume replaces it, mirroring `'compressing'`. Unlike manual `/compress` it does **not** also push a transcript line, since automatic compaction can fire on many turns of a large session and would be noisy. Existing `'compressing'` behavior is unchanged.

## Tests
- New event-handler tests (fake timers): `'compacting'` stays pinned past 4s and past 120s; a later status event replaces it on resume; no transcript line is pushed for automatic compaction.
- `vitest run src/__tests__/createGatewayEventHandler.test.ts` -> 92 passed. Mutation check: disabling the new branch makes the >4s test fail (status reverts to `ready`), confirming the fix is exercised.
- Verified on Ubuntu 24.04.

PR #98371, opened 2026-08-30
compression: emit periodic client-visible status during compaction

JackHunzicker:
## Problem

Context compression can run for many minutes with **no deltas, tool events, or status lines** reaching remote transports. Clients that run idle-progress turn watchdogs treat that silence as a dead turn and kill it — the Android relay app fires `session.interrupt` after 180s of no events (`GatewayChatClient.kt` `TURN_TIMEOUT_MS = 180_000L`).

On a session near the context ceiling this becomes an **infinite loop**: every prompt triggers preflight compression → compression is silent → the client watchdog interrupts at exactly +180s → the work rolls back → the next prompt retriggers it. The session becomes permanently unresponsive from that client.

Observed in production tonight (gpt-5.6-sol via openai-codex, whose backend window recently dropped to 272k, making preflight compression far more common):

```
compression attempt telemetry: {"commit_status":"aborted","failure_class":"explicit_interrupt","total_duration_ms":180469}
compression attempt telemetry: {"commit_status":"aborted","failure_class":"explicit_interrupt","total_duration_ms":180233}
compression attempt telemetry: {"commit_status":"aborted","failure_class":"explicit_interrupt","total_duration_ms":180219}
Turn ended: reason=interrupted_by_user  (four times, each at prompt+~251s = visible tool work + exactly 180s of silent compression)
```

## Fix

`_CompressionActivityHeartbeat` already ticks every 60s during compression, but only refreshes the internal SessionDB activity tracker. This PR makes it also emit a client-visible `compacting` status through the existing `agent.status_callback` seam — once at compression start and on every tick. The gateway already routes `status_callback` → `status.update` events (`_status_update` even has a `compacting` re-tag path), and clients already reset their watchdogs on any received gateway event, so each heartbeat re-arms them.

No new config, no new plumbing — one method on the existing heartbeat class plus two call sites.

## Testing

- New test `test_compression_activity_heartbeat_emits_client_status_events` (mirrors the existing heartbeat test): asserts ≥2 `compacting` status events during a slow compress (start + at least one tick).
- Full file: `tests/agent/test_compression_concurrent_fork.py` — **47 passed** (all pre-existing tests unaffected; the completion-edge assertions use `.count()`/`not in`, so additive `compacting` events don't disturb them).

Companion client-side PR (defensive longer watchdog leash on `compacting` status): Codename-11/hermes-relay.

### DIFF
--- a/agent/conversation_compression.py
+++ b/agent/conversation_compression.py
@@ -1857,6 +1857,7 @@ class _CompressionActivityHeartbeat:
         # if a prior timeout/cooldown stamp is still on the agent.
         self._suppressed = False
         self._touch("context compression started", allow_terminal_overwrite=True)
+        self._emit_progress_status()
         self._thread.start()
         return self
 
@@ -1917,11 +1918,38 @@ class _CompressionActivityHeartbeat:
         except Exception:
             logger.debug("compression activity heartbeat touch failed", exc_info=True)
 
+    def _emit_progress_status(self) -> None:
+        """Publish a client-visible ``compacting`` status line.
+
+        Compression can stream for minutes with no deltas, tool events, or
+        status lines reaching remote transports. Idle-progress watchdogs on
+        those clients (e.g. the Android relay app's 180s turn watchdog)
+        treat the silence as a dead turn and fire ``session.interrupt`` —
+        killing a healthy compression mid-flight and rolling back its work,
+        which retriggers on the next prompt and loops forever on sessions
+        near the context ceiling. A periodic ``status_callback`` heartbeat
+        gives every transport a progress event to re-arm on.
+        """
+        status_callback = getattr(self._agent, "status_callback", None)
+        if not status_callback:
+            return
+        try:
+            status_callback(
+                "compacting",
+                f"🗜️ {COMPACTION_STATUS_MARKER} — still summarizing "
+                "earlier conversation so I can continue...",
+            )
+        except Exception:
+            logger.debug(
+                "status_callback error in compression heartbeat", exc_info=True
+            )
+
     def _run(self) -> None:
         while not self._stop.wait(self._interval_seconds):
             if self._should_suppress():
                 return
             self._touch("context compression in progress")
+            self._emit_progress_status()
 
 def _direct_messages_for_pre_compress_memory(messages: Any) -> list[dict[str, Any]]:
     """Return direct user/assistant evidence safe for memory checkpointing.
--- a/tests/agent/test_compression_concurrent_fork.py
+++ b/tests/agent/test_compression_concurrent_fork.py
@@ -167,6 +167,48 @@ def test_compression_activity_heartbeat_touches_agent_during_long_compress(tmp_p
     assert db.get_compression_lock_holder(session_id) is None
 
 
+def test_compression_activity_heartbeat_emits_client_status_events(tmp_path: Path) -> None:
+    """The heartbeat must emit ``compacting`` status events, not just DB touches.
+
+    Remote transports (e.g. the Android relay app) run idle-progress turn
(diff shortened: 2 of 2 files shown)


alt-glitch:

Related to #88102 and #70365's compaction-status work. This PR addresses the distinct remote-watchdog interruption mechanism by emitting periodic client-visible status events.

liyangbing:
<h3>Review</h3><p>The client-visible heartbeat targets the right failure mode: compression can be healthy while remote clients see silence and interrupt it. The remaining contract is to make a status event a liveness signal without allowing the reporting path to extend, corrupt, or falsely complete the compression commit.</p><p>Please make these invariants explicit:</p><p>1. Emit an initial status only after compression is actually admitted, and stop emitting promptly on success, rollback, cancellation, exception, or timeout.<br>2. Keep heartbeat delivery non-blocking or bounded; a slow/failing status callback must never hold the compression fence or change the commit result.<br>3. Include session/profile identity, compression generation, elapsed time, and phase so delayed events from an old attempt cannot reset a client watchdog for a newer turn.<br>4. Define cadence and client semantics: repeated <code>compacting</code> events refresh liveness, but a missing heartbeat must not be treated as proof that the server is dead, and completion must supersede queued heartbeats.<br>5. Preserve compatibility for clients that do not understand the new status and ensure the status path does not leak prompt, tool, or provider data.</p><h3>Regression matrix</h3><p>Please cover:<br>• start, periodic tick, successful finish, rollback, explicit interrupt, exception, and commit timeout<br>• slow status callback, callback exception, gateway backpressure, disconnect, reconnect, and client without the new status<br>• delayed/out-of-order heartbeat after completion or a newer compression generation<br>• multiple concurrent Sessions/profiles with isolated status identity<br>• heartbeat cadence around the remote 180-second watchdog and long compression exceeding several ticks<br>• durable compression lock/fence release and transcript integrity when status delivery fails</p><p>Direct evidence: <a href="https://github.com/NousResearch/hermes-agent/pull/98371">PR #98371</a>, which changes <code>agent/conversation_compression.py</code> and adds the heartbeat status regression test.</p>

teknium1:
Thanks for this — the remote-watchdog interruption loop you documented (client `session.interrupt` after 180s of event silence killing a healthy compression) is a real and distinct failure mode from the TUI-visibility half of #97239.

For the TUI-visibility cluster we salvaged @HexLab98's #97253 (earliest submitter, root-cause retag fix) onto current main as PR #99512; you're credited in its description.

Why not this heartbeat as the primary fix: it emits new status wording (`"— still summarizing earlier conversation…"`) that isn't one of the routine template constants, so under the `compression.progress_notices` gate architecture it would need the constant + `ROUTINE_COMPRESSION_STATUS_SAMPLES` + noise-regex trio treatment to avoid either leaking to chat platforms or being silently filtered. The periodic-heartbeat idea for re-arming client watchdogs remains worth pursuing on its own — leaving this open for that discussion and Teknium's call as merge arbiter.

kshitijk4poor:
Salvaged via #101733 (merged; your commit cherry-picked with authorship preserved) + #101903. On @teknium1's bar: the heartbeat is now a registered constant (`COMPACTION_HEARTBEAT_STATUS`), emitted under the `lifecycle` key through `agent._emit_status`, in `ROUTINE_COMPRESSION_STATUS_SAMPLES`, and both gateway regexes derive from the constants — so it is filtered on chat platforms exactly like `COMPACTION_STATUS` and re-tagged to `compacting` for the TUI/desktop. Also gated on the quiet-engine contract (no visible phase without a terminal edge) and the duplicate start line dropped. Thanks @JackHunzicker for documenting the watchdog loop precisely. Closing in favour of the salvage PR.

PR #98372, opened 2026-08-30
test(cli): cover _startup_fast probe and dispatch behavior

salch-cred:
Add behavioral regression coverage for `hermes_cli._startup_fast` — the pre-import fast-path helpers.

## Why

`_startup_fast` holds the canonical pre-import logic (project-root resolution, Termux detection, container probing, active-profile home override, OpenAI version parsing, install-method stamp, and the `--version` fast-path dispatch). It runs before the heavy import wall in `main.py`, so a regression here can break `hermes --version` on a whole environment class (the exact drift that killed the Termux fast path). The existing test file covers only import-weight and version parity — the actual probe/dispatch behavior had no direct coverage.

## What

`tests/hermes_cli/test_startup_fast_behavior.py` — behavioral tests:

- `is_termux_env` via `TERMUX_VERSION` and via the `com.termux` prefix, and the false case.
- Version argv acceptance (`--version`/`-V`) and rejection of other args.
- Container detection via `/.dockerenv`, via cgroup markers (`/docker/`, `/podman/`, `/lxc/`), and the no-marker false case.
- `active_profile_may_override_home` for a non-default profile, `default`, and a missing marker.
- `read_openai_version` parses the SDK version; returns None when absent.
- `read_install_method` reads the `.install_method` stamp; None when missing.
- `try_fast_version` dispatches `--version`, skips when container mode may be active, and honors the Termux escape hatch.
- `project_root_str` resolves to the repo root.

## Validation

All assertions verified to pass against the current `_startup_fast.py` on `main`.

### DIFF
--- a/tests/hermes_cli/test_startup_fast_behavior.py
+++ b/tests/hermes_cli/test_startup_fast_behavior.py
@@ -1,0 +1,167 @@
+"""Behavioral tests for hermes_cli._startup_fast — the pre-import fast-path helpers.
+
+The existing tests cover import-weight and version parity. These pin the actual
+behavior of the pure probe/dispatch functions so a subtle env/path regression
+(e.g. the drift that killed the Termux fast path) can't slip through.
+"""
+
+import io
+
+import pytest
+
+import hermes_cli._startup_fast as sf
+
+
+# ── is_termux_env ─────────────────────────────────────────────────────────────
+
+def test_is_termux_env_via_TERMUX_VERSION(monkeypatch):
+    monkeypatch.setenv("TERMUX_VERSION", "0.118.1")
+    monkeypatch.delenv("PREFIX", raising=False)
+    assert sf.is_termux_env() is True
+
+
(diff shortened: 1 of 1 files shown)


Enough1122:

**Overall:** Test-only addition covering hermes_cli._startup_fast pure helpers.

**What it does**
- New `tests/hermes_cli/test_startup_fast_behavior.py` adds ~15 tests: `is_termux_env` via TERMUX_VERSION/PREFIX, version argv dispatch (`--version`/`-V` exact match), container detection via dockerenv/cgroup, active_profile override (non-default vs default vs missing), `read_openai_version` parsing, `read_install_method` stamp, `try_fast_version` dispatch including container-active skip and Termux disable hatch, and `project_root_str`.

**Non-blocking notes**
- Tests use monkeypatch on `sf.os.path.exists` and `builtins.open` — isolated and no real FS dependency.

No issues — pure coverage improvement.

PR #51519, opened 2026-06-23
fix(gateway): scope idempotency fingerprint to session/memory context

Que0x:
## What & why
The Idempotency-Key cache fingerprinted only selected JSON body fields,
ignoring request context that lives outside the body. Two requests with
the same key and body but different `X-Hermes-Session-Id`,
`X-Hermes-Session-Key`, or explicit `conversation_history` produced the
same fingerprint — so the second caller got the first's cached response,
computed under a different session / memory / history context.

## Fix
- **Chat:** fold the effective `session_id` and `gateway_session_key`
  into the fingerprint.
- **Responses:** add `conversation_history` + `truncation` to the keys
  and `gateway_session_key` as context. `session_id` is deliberately
  excluded — it's a fresh UUID per new response and would break
  legitimate retries; its chaining is already covered by
  `previous_response_id` / `conversation`.
- `_make_request_fingerprint` gains an optional `extra` arg; omitting it
  is byte-for-byte backward compatible.

Genuine duplicates (same key, body, and context) still dedupe.

## Tests
Added `TestRequestFingerprint` (4 unit) and `TestIdempotencyRequestContext`
(5 endpoint) regression tests. Confirmed the "runs twice" cases fail on the
pre-fix code (agent called once) and pass after.

`pytest tests/gateway/test_api_server.py` → **174 passed, 1 skipped**.

PR #36183, opened 2026-06-01
fix: add threading.Lock to ResponseStore to prevent FD leak

HymanZ:
## Problem

SQLite connections opened with `check_same_thread=False` create new file descriptors when accessed from concurrent threads. Under the Hermes cron scheduler (ThreadPoolExecutor) combined with API server request threads, all hitting the same `ResponseStore._conn` without serialization, this produces duplicate `.db` and `.db-wal` handles — **109 `.db` + 108 `.db-wal` observed** before hitting the ulimit.

**Symptom**: cron jobs silently fail when total FD count exceeds the system ulimit (256 by default on macOS). Error surfaces only in gateway logs, not to the scheduler — so cron jobs fail silently for hours.

## Fix

Mirror `SessionDB`'s approach in `hermes_state.py`: wrap all `ResponseStore` methods with a `threading.Lock` so the single `sqlite3.Connection` is only ever touched by one thread at a time, preventing the WAL/SHM descriptor duplication.

- Same file: `gateway/platforms/api_server.py`
- No external API changes — all insertions are internal to the ResponseStore class
- No behavioral change for single-threaded access

## Relationship to #37660 (merged official fix)

The official fix in #37660 (`close ResponseStore + dispose unowned adapter on reconnect failure`) addresses a **different root cause**: orphaned adapters accumulating from failed reconnect loops. This PR addresses **concurrent thread access** creating duplicate SQLite handles. They complement each other:

| Mechanism | #37660 (merged) | This PR (#36183) |
|-----------|----------------|-------------------|
| Root cause | Reconnect failure leaves orphan adapters | Concurrent threads create duplicate handles |
| Layer | Cleanup on failure | Serialization at access |
| Effect | Prevents 12h-exhaustion from retry loops | Prevents concurrent FD explosion |
| Risk profile | Zero — only adds cleanup | Zero — proven pattern from SessionDB |

Without this PR, #37660 cleans up orphaned connections on reconnect failure but does not prevent WAL/SHM descriptor duplication when multiple threads access the active connection concurrently.

## Verification

After the fix, `response_store.db` FD count stable at **3** (1 .db + 1 .db-wal + 1 .db-shm), no longer growing with thread count.

## Review Status

✅ Approved by 2 reviewers (mxnstrexgl, tonydwb)
📌 Blocked on CI — fork PR does not trigger CI pipeline

PR #7578, opened 2026-04-11
fix(gateway): add threading.Lock to ResponseStore for concurrent access safety

chinadbo:
## Summary
- Add `threading.Lock` to `ResponseStore` class wrapping all public methods (`get`, `put`, `delete`, `get_conversation`, `set_conversation`, `__len__`)
- Prevents TOCTOU logic bugs in LRU eviction counting and access-time tracking under multi-threaded gateway execution
- Without locking, concurrent puts/gets can cause `SystemError` from sqlite3 when `commit()` is called concurrently

## Test plan
- [x] New tests in `tests/gateway/test_response_store_threading.py` verify concurrent put does not over-evict, concurrent get/put does not crash, and LRU access-time updates are atomic
- [x] Existing `TestResponseStore` tests (7 tests) all pass with no regression
- [x] No secrets, API keys, or sensitive data in changes

PR #83546, opened 2026-08-11
refactor(gateway): extract api_server idempotency cluster to dedicated module

andrexibiza:
refactor(gateway): extract api_server idempotency cluster to dedicated module

Part of #78647
Part of #78643

## What

Byte-verbatim extraction of the idempotency cluster (`_IdempotencyCache`,
`_idem_cache`, `_make_request_fingerprint`; gateway/platforms/api_server.py
lines 1210–1261 at pin `ee4bb75b532`) into a new module
`gateway/platforms/api_server_idempotency.py`, with an identity-preserving
three-name re-export shim in api_server.py. Zero behavior change.

## Method (5×2×3 double-blind)

- Wave 1: blind region analyst (R1) — six-artifact deliverable on disk.
- Wave 2: blind adversarial witness (R1) — independent replication with
 adversarial prior.
- Wave 3: consensus adjudicator — re-verified every load-bearing claim at the
 pin; applied the fixed tiebreak order. w2's window 149–409 was BLOCKED by
 the live open-PR gate (#65009/#66346/#80587 definitive collisions); w1's
 idempotency window 1210–1261 selected. Census correction recorded:
 `tests/gateway/test_api_server.py:33` imports `_IdempotencyCache` → the
 re-export must carry all three names.
- Blind implementer: executed the consensus contract only; byte-verbatim
 move, golden sha verified pre/post.
- Two new blind reviewers (correctness + adversarial), mutually blind:
 **both APPROVED** with direct execution evidence.

## Evidence

- Golden sha (window bytes at pin): `40519eacf19bb21ac7fc3afe348249ffb0451123a00a88abb563a9df758f04f6` — re-derived by implementer, both reviewers, and the adjudicator.
- Seam identity: all three names resolve to identical objects through `gateway.platforms.api_server` (live import checks).
- Singleton correctness: `_idem_cache` is the SAME object through both modules (no second instance).
- Internal call sites (4279/4284/5404/5418) keep resolving through the re-export; `test_api_server.py:33/:126` import path intact.
- Scope: api_server.py −52/+1 (shim), +api_server_idempotency.py, +seam test.

## Coordination

- Sibling surfaces: watch items #66283/#75519/#78467 (idempotency-semantic; diff-check before landing per contract §5.4).
- Dependencies: none. Merge order: independent.
- Dedup: no prior or concurrent PR extracts this cluster (verified against open PR census 2026-08-10).
[...]

PR #98373, opened 2026-08-30
fix(api_server): synchronize ResponseStore access, use atomic subquery eviction, and sort fingerprint keys

ericmaddox:
## Summary
This PR synchronizes `ResponseStore` access across threads/coroutines, hardens server teardown lifecycle ordering, replaces chunked dynamic-variable SQL deletion with atomic subquery LRU eviction, and guarantees deterministic and session-scoped request fingerprinting in `api_server.py`.

### Key Improvements & Invariants
1. **Thread & Coroutine Synchronization with Terminal Fence**:
   - Protects all SQLite operations in `ResponseStore` (`get`, `put`, `delete`, `get_conversation`, `set_conversation`, `__len__`, `close`) using an internal `threading.RLock()`.
   - Adds a terminal `_closed` boolean guard so post-close access safely no-ops or returns `None` rather than raising `sqlite3.ProgrammingError`.
   - Aligns `APIServerAdapter.disconnect()` teardown order to quiesce and stop `_site` and drain `_runner` before closing `_response_store`.

2. **Atomic Single-Pass Subquery Eviction**:
   - Replaces `SELECT id ...` followed by `DELETE FROM ... WHERE id IN (?, ?, ...)` with single-statement atomic subqueries:
     ```sql
     DELETE FROM conversations WHERE response_id IN (
         SELECT id FROM responses ORDER BY created_at ASC, id ASC LIMIT ?
     );
     DELETE FROM responses WHERE id IN (
         SELECT id FROM responses ORDER BY created_at ASC, id ASC LIMIT ?
     );
     ```
   - Completely avoids SQLite's variable limit (`SQLITE_MAX_VARIABLE_NUMBER`, default 999) when excess items > 999.

3. **Deterministic & Session-Scoped Request Fingerprinting**:
   - Canonicalizes JSON serialization via `json.dumps(..., sort_keys=True)` to prevent hash divergence due to dict key ordering.
   - Binds effective `session_id`, `gateway_session_key`, and `extra_context` into the fingerprint hash to guarantee requests in different session or memory scopes never collide.

4. **Comprehensive Regression Suite**:
   - Verified with 1200-item single-pass eviction exceeding SQLite variable limits.
   - Tested terminal `_closed` store fence, concurrent multi-threaded store access, and session-scoped / key-ordered request fingerprinting.

Closes #7578, closes #36183, closes #51519.

Co-authored-by: chinadbo <chinadbo@users.noreply.github.com>
Co-authored-by: HymanZ <HymanZ@users.noreply.github.com>
Co-authored-by: Que0x <Que0x@users.noreply.github.com>

### DIFF
--- a/gateway/platforms/api_server.py
+++ b/gateway/platforms/api_server.py
@@ -660,6 +660,8 @@ class ResponseStore:
 
     def __init__(self, max_size: int = MAX_STORED_RESPONSES, db_path: str = None):
         self._max_size = max_size
+        self._lock = threading.RLock()
+        self._closed = False
         if db_path is None:
             db_path = ":memory:"
             with suppress(Exception):
@@ -674,13 +676,26 @@ class ResponseStore:
         # Shared WAL-fallback so response_store.db degrades gracefully on NFS/SMB/FUSE homes.
         from hermes_state_wal import apply_wal_with_fallback
         apply_wal_with_fallback(self._conn, db_label="response_store.db")
-        self._conn.execute(
-            "CREATE TABLE IF NOT EXISTS responses ("
-            "response_id TEXT PRIMARY KEY, data TEXT NOT NULL, accessed_at REAL NOT NULL)")
-        self._conn.execute(
-            "CREATE TABLE IF NOT EXISTS conversations (name TEXT PRIMARY KEY, response_id TEXT NOT NULL)")
-        self._conn.commit()
-        # Conversation history lives here: owner-only perms, once at init (not per commit).
+        with self._lock:
+            self._conn.execute(
+                """CREATE TABLE IF NOT EXISTS responses (
+                    response_id TEXT PRIMARY KEY,
+                    data TEXT NOT NULL,
+                    accessed_at REAL NOT NULL
+                )"""
+            )
+            self._conn.execute(
+                """CREATE TABLE IF NOT EXISTS conversations (
+                    name TEXT PRIMARY KEY,
+                    response_id TEXT NOT NULL
+                )"""
+            )
+            self._conn.commit()
+        # response_store.db contains conversation history (tool payloads,
+        # prompts, results). Tighten to owner-only after creation so other
+        # local users on a shared box can't read it. Run once at __init__
+        # rather than after every commit — chmod-on-every-write is wasted
+        # syscalls on a hot path.
         self._tighten_file_permissions()
 
     def _tighten_file_permissions(self) -> None:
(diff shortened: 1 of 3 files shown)


alt-glitch:

Related to #7578 and #36183: all guard ResponseStore's shared SQLite connection. This PR also adds distinct eviction and idempotency-fingerprint fixes; reviewers should choose or consolidate the locking approach.

andrexibiza (commented):
I like the direction here: serializing the shared `sqlite3.Connection` is the right primitive, the subquery eviction removes the dynamic-placeholder ceiling, and canonical JSON is strictly better than insertion-ordered `repr()` for idempotency identity. I checked the exact current head `006d33f8f1ec3bfd100f6ef021dc1477357820b9`, the two-file diff, the full `api_server.py` lifecycle around `ResponseStore`, the new tests, current `main`, exact-head workflows, and the prior/adjacent ownership graph.

Exact-head hosted evidence is green: [CI 33294976548](https://github.com/NousResearch/hermes-agent/actions/runs/33294976548), [Docker 33294976044](https://github.com/NousResearch/hermes-agent/actions/runs/33294976044), and [Nix 33294976025](https://github.com/NousResearch/hermes-agent/actions/runs/33294976025) all succeeded; CI includes successful contributor attribution. There is only one submitted commit, so the current submitted commit is green.

I do not think this exact object is landing-ready yet. Two P1 boundaries and one verification/provenance gate remain.

### P1 — the lock serializes `close()`, but the adapter still closes the store before HTTP work is quiesced

At this head, `ResponseStore.close()` takes `_lock` and closes `self._conn`, but there is no terminal/closed-state fence. Separately, `APIServerAdapter.disconnect()` still does this in the opposite lifetime order:

1. `self._response_store.close()`
2. then `await self._site.stop()`
3. then `await self._runner.cleanup()`

That means the new lock prevents an in-flight SQL statement from overlapping the close, but it does **not** prevent a request that was already admitted by aiohttp from acquiring the lock *after* the close and executing `self._conn.execute(...)` on a closed connection. The failure becomes a deterministic `sqlite3.ProgrammingError` use-after-close instead of a concurrent cursor race.

This is not a hypothetical sibling shape. Merged #37679 deliberately made reconnect/shutdown disposal close `ResponseStore`, so this lifecycle is now part of the live reconnect pressure path. Open #7578 by `chinadbo` already contains the missing half of the shape: it adds a `_closed` terminal guard and a concurrent reader-vs-close regression in addition to the same serialization primitive. #98373 currently reimplements the locking owner but drops that shutdown-safety behavior.

**Required fix:** establish one terminal lifetime boundary for the store. Prefer quiescing/stopping the HTTP site and draining the runner before closing the DB, and keep a defensive closed-state/`None` fence so no public store method can touch a retired connection. Add a regression that holds a real request or store operation across `disconnect()` and proves no post-close DB access occurs. The synchronization contract needs to cover `active → quiescing → closed`, not only concurrent SQL calls.

### P1 — `_make_request_fingerprint()` is now deterministic, but still under-scoped to the body

The new serializer makes a selected body subset canonical, but the exact-head helper still has the shape:

```python
def _make_request_fingerprint(body, keys):
    subset = {k: body.get(k) for k in keys}
    ... json.dumps(subset, sort_keys=True) ...
```

The response is not a function of that body subset alone. The chat path also consumes the effective `X-Hermes-Session-Id` / session history and `X-Hermes-Session-Key` memory scope; the Responses path has equivalent response-affecting conversation/history/truncation context. Reusing the same `Idempotency-Key` + body under a different session/memory scope therefore still produces the same fingerprint and can return the first cached response under the wrong context.

Open #51519 by `Que0x` is the current direct implementation of this missing scope boundary; closed-unmerged #24200 is earlier independent security provenance for the same class. #51519 adds the effective session/memory context for chat and the corresponding Responses context, with endpoint regressions proving same-key/same-body requests in different scopes execute independently.

Because #98373 is changing the canonical fingerprint owner, please compose this rather than landing a stronger serialization of an identity that is still missing response-affecting scope. **Required fix:** fingerprint the complete semantic request identity: canonicalized selected body plus the trusted effective session/memory/history context actually consumed by `_run_agent`, and preserve the #51519/#24200 lineage rather than silently superseding it.

### Verification blocker — the “>1000-item eviction” test never creates a >1000-item excess

`test_large_excess_eviction_exceeding_sql_variable_limit` constructs `ResponseStore(max_size=5)` and then calls `store.put(...)` 1200 times. But every `put()` immediately prunes back to 5. On each iteration the pre-eviction count is at most 6, so `count - max_size` is at most **1**. The test never builds an `evict_ids` list larger than one item on the old implementation and therefore cannot exercise the claimed SQLite variable-limit failure.

Please make the regression red on the old code: e.g. populate a persistent DB under a much larger `max_size`, close it, reopen the same DB with a much smaller `max_size`, then perform one `put()` so the single eviction pass has >999 victims; or seed the oversized rows directly. Verify conversation mappings for the evicted rows as well. That turns this from a plausible optimization into a witnessed bug closure.

### Ownership / merge order / attribution

There are three direct collisions that should be resolved as one convergence, not three independent owners:

- #7578 (`chinadbo`) — earlier `ResponseStore` serialization plus the unique close/late-reader fence and threading regressions.
- #36183 (`HymanZ`) — independent `ResponseStore` serialization with field evidence around the same shared-connection failure class.
- #51519 (`Que0x`) — current semantic-scope fix for the same `_make_request_fingerprint` helper and the same `tests/gateway/test_api_server.py` surface.

#98373 has useful unique work — the subquery eviction and canonical JSON serializer — so I would treat it as a **superseding convergence candidate**, not a duplicate. But the convergence object should preserve those contributors' provenance and the stronger behaviors/tests already discovered by the earlier carriers.

There is also a structural collision with open #83546, the idempotency-cluster extraction (`_IdempotencyCache`, `_idem_cache`, `_make_request_fingerprint`) out of `api_server.py`. That branch is stale/non-mergeable today, so it is not a landing object, but its ownership direction matters: settle the final idempotency semantics first, then refresh/extract that final owner; do not merge a byte-verbatim older extraction after #98373 and accidentally restore pre-fix semantics.

Finally, current `main` is `26350357d76e4508c8df9304a3374bdc5a6f6220`, while this commit is parented by merge-base `52e5e7c034edfbd0105d0b9c315c4eecb6be7f5e`; GitHub compare reports this head **1 ahead / 17 behind**. It is mergeable, but the repaired convergence object should be rematerialized on current main and reacquire fresh exact-head CI/Docker/Nix. The current green receipts are valid for `006d33f…`; they do not transfer to a rebased/recomposed commit.

The core work is worth keeping. The key is to make the synchronization survive teardown, make the fingerprint carry the whole response-affecting identity, and make the eviction regression actually cross the boundary it claims to close. Once those are composed with the prior owners, this can close several long-running API-server defect threads at once. 🚀

liyangbing:
<h3>Review</h3><p>The PR addresses three independent failure modes in a shared SQLite-backed <code>ResponseStore</code>. The important correctness boundary is wider than protecting individual method bodies: every read/modify/write sequence, commit/rollback, and close must have one clearly owned transaction boundary when async handlers and background threads share the connection.</p><p>Please make these invariants explicit:</p><p>1. Hold the store lock across cursor creation, all rows consumed, commit/rollback, and error cleanup; a cursor or transaction must not escape the protected region.<br>2. Define <code>close()</code> versus in-flight operations: close should either wait for admitted work or reject new work deterministically, and subsequent methods must return a stable closed-store error rather than a partially initialized connection failure.<br>3. Keep eviction atomic with insertion and preserve deterministic LRU order under ties; verify that an interrupted transaction cannot leave the newest response absent or evict a still-referenced conversation.<br>4. The key-order-invariant fingerprint should also define treatment of omitted versus explicit <code>null</code>, lists, non-JSON values converted by <code>default=str</code>, and fields that must never participate in idempotency. Otherwise two semantically different requests may collide.<br>5. Ensure the synchronization strategy remains valid when separate processes open the same database; an in-process <code>RLock</code> prevents thread races but does not replace SQLite busy-timeout/transaction handling across processes.</p><h3>Regression matrix</h3><p>Please cover:<br>• concurrent get/put/delete/conversation operations with forced interleavings<br>• close racing with each public operation, including repeated close and post-close calls<br>• insert plus eviction failure, rollback, retry, and deterministic tie ordering<br>• more than SQLite’s variable limit with mixed referenced/unreferenced and terminal/non-terminal rows<br>• two processes sharing the database, busy/locked responses, and recovery after a writer exits<br>• fingerprints with reordered keys, omitted versus <code>null</code>, nested lists/maps, unusual scalar values, and excluded volatile fields<br>• duplicate request retry after a partial failure, proving the idempotency result is neither lost nor incorrectly reused</p><p>Direct evidence: <a href="https://github.com/NousResearch/hermes-agent/pull/98373">PR #98373</a>, which changes <code>gateway/platforms/api_server.py</code> and adds <code>tests/gateway/test_api_server.py</code> coverage for the shared store and fingerprint behavior.</p>

ericmaddox:
Thank you for the thorough and constructive review, @andrexibiza!

All requested changes and convergence requirements have been implemented and pushed in `13520e3`:

### 1. Teardown Ordering & Terminal `_closed` Fence (P1)
- **Quiesced Teardown**: In `APIServerAdapter.disconnect()`, `await self._site.stop()` and `await self._runner.cleanup()` are now awaited *before* `self._response_store.close()` runs in the `finally` block, ensuring in-flight aiohttp handlers drain completely before database disposal.
- **Terminal Closed Fence**: Added `self._closed = True` on `ResponseStore`. All public methods (`get`, `put`, `delete`, `get_conversation`, `set_conversation`, `__len__`) check `if self._closed or self._conn is None:` under `self._lock` and deterministically return neutral values / safely no-op.
- **Test Coverage**: Added `test_response_store_closed_fence` to verify that operations after `close()` safely no-op without `sqlite3.ProgrammingError` and that repeated `close()` is idempotent.

### 2. Request Fingerprint Semantic Scoping (P1)
- **Scope Expansion**: Updated `_make_request_fingerprint(body, keys, session_id=..., gateway_session_key=..., extra_context=...)` to incorporate the effective `session_id`, `gateway_session_key`, and any additional context into the canonicalized key subset.
- **Handler Updates**: Wired resolved `session_id` and `gateway_session_key` into `_make_request_fingerprint` calls in both `_handle_chat_completions` and `_handle_responses`.
- **Test Coverage**: Added `test_request_fingerprint_scoped_to_session_and_memory` verifying identical payloads under distinct session/memory contexts produce independent fingerprints.

### 3. Eviction Boundary Verification
- **True Excess Test**: Rewrote `test_large_excess_eviction_exceeding_sql_variable_limit(self, tmp_path)` to pre-populate 1,200 items into a persistent SQLite DB under `max_size=2000`, close it, and reopen with `max_size=5`. A single `put()` call now forces a single-pass atomic eviction of 1,196 items, proving immunity to SQLite's variable limit (>999).

### 4. Rebase & Attribution Convergence
- Rebased cleanly onto the latest `upstream/main` (`63279301bc`).
- Credited co-authors in commit trailers to preserve provenance for the converged fixes:
  - `Co-authored-by: chinadbo <chinadbo@users.noreply.github.com>` (#7578)
  - `Co-authored-by: HymanZ <HymanZ@users.noreply.github.com>` (#36183)
  - `Co-authored-by: Que0x <Que0x@users.noreply.github.com>` (#51519)

All gateway test suites pass locally (16/16 selected tests in `test_platform_reconnect_fd_leak.py` and `test_api_server.py`). Fresh CI run is now underway.

PR #98375, opened 2026-08-30
fix(backup): reject -o/--output with --quick instead of silently drop…

### LINKED ISSUE #98369: [Bug]: hermes backup --quick silently ignores -o/--output — exits 0, file is never created
## Bug Description

`hermes backup` accepts `-o/--output`. With `--quick`, the flag is parsed, accepted, and
then never read: no file is written at the requested path, no warning is printed, and the
command exits `0`.

The dispatch is two lines (`hermes_cli/main.py:5914`):

```python
def cmd_backup(args):
    if getattr(args, "quick", False):
        run_quick_backup(args)      # reads args.label only
    else:
        run_backup(args)            # honours args.output
```

`run_quick_backup` (`hermes_cli/backup.py:2064`) never touches `args.output`:

```python
def run_quick_backup(args) -> None:
    """CLI entry point for hermes backup --quick."""
    label = getattr(args, "label", None)
    snap_id = create_quick_snapshot(label=label)
    ...
```

while the non-quick branch does (`hermes_cli/backup.py:801`):

```python
if args.output:
    out_path = Path(args.output).expanduser().resolve()
```

So the same flag on the same command is honoured in one branch and silently dropped in the
other.

The parser already knows how to express "this option only applies to one branch" — it does
it for `--label` (`hermes_cli/subcommands/backup.py:35`):

```python
backup_parser.add_argument(
    "-l", "--label", help="Label for the snapshot (only used with --quick)"
)
```

`--output` has no such note, no `parser.error()`, and no runtime warning.

## Steps to Reproduce

Deterministic, in a throwaway container from the released image
(`nousresearch/hermes-agent:v2026.8.27`) — no gateway, no network, no existing install:

```sh
export HERMES_HOME=/tmp/hh
P=/opt/hermes/.venv/bin/python
mkdir -p "$HERMES_HOME"
printf "model:\n  default: gpt-4\n" > "$HERMES_HOME/config.yaml"
"$P" -c "import sqlite3; c=sqlite3.connect('$HERMES_HOME/state.db'); c.execute('CREATE TABLE t (a)'); c.commit()"
cd /opt/hermes

"$P" -m hermes_cli.main backup --quick -o /tmp/requested.zip ; echo "exit=$?"
ls -la /tmp/requested.zip

"$P" -m hermes_cli.main backup -o /tmp/full.zip
ls -la /tmp/full.zip
```

Actual output:

```
### 1. quick backup, asking for a specific output path
State snapshot created: 20260830-052437
  1 snapshot(s) stored in /tmp/hh/state-snapshots/
  Restore with: /snapshot restore 20260830-052437
exit=0
-- is /tmp/requested.zip there?
ls: cannot access '/tmp/requested.zip': No such file or directory

### 2. same command, same flag, without --quick
Restore with: hermes import full.zip
-- is /tmp/full.zip there?
-rw-r--r-- 1 1158 Aug 30 05:24 /tmp/full.zip
```

The same check without running anything:

```
>>> import inspect, hermes_cli.backup as b
>>> "output" in inspect.getsource(b.run_quick_backup)
False
```

## Expected Behavior

Any one of these closes the gap; which one is the maintainers' call:

- `--quick` honours `-o` and writes the snapshot where asked;
- the parser rejects the combination (`parser.error`);
- `--output`'s help says it does not apply to `--quick`, the way `--label`'s help already
  says the opposite.

What should not happen is the current outcome: a requested path, exit `0`, and no file.

## Impact

This is easy to miss precisely because the command succeeds. Any caller that records "the
backup is at `<the path I asked for>`" records a pointer to a file that was never created,
and no return code or output check catches it — the snapshot really was made, just
somewhere else, under an id the caller never sees unless it parses stdout.

It matters most for automation around restores, where the recorded path is what a restore
would later read.

## Prior art

[PR #47469](https://github.com/NousResearch/hermes-agent/pull/47469) (June 2026) changed
this, and described it in the same terms:

> `hermes_cli/backup.py`: Wire up the `-o`/`--output` flag in `run_quick_backup` to write a
> `.tar.gz` of the snapshot directory — previously the parser accepted `-o` for `--quick`
> but silently ignored it

It was closed by its author, with no review and no comments. We mention it only so the
history is visible: the behaviour has been noticed before, nothing was decided about it,
and the code is unchanged today. That PR also bundled two other changes (adding
`skills`/`plugins`/`scripts` to the snapshot, and a directory-walking rewrite), so it was
not a single-purpose change.

## Version

  release at the time of writing, and the one the reproduction above was run in.
- Still present on `main`: `run_quick_backup` is unchanged at `hermes_cli/backup.py:2064`.

```sh
hermes backup --quick -o /tmp/x.zip && ls /tmp/x.zip
```

If the command succeeds and `ls` fails, it is back.

bsdnn:
## What does this PR do?

`hermes backup --quick` accepted `-o/--output`, then never read it:
`run_quick_backup` stores a state snapshot into `HERMES_HOME/state-snapshots/`
and never writes an archive, so the requested path was silently ignored — the
command exited 0 with **no file created**, while the non-quick branch honored
the same flag (#98369).

Fix: make `-o/--output` and `-q/--quick` a mutually exclusive argparse group, so
the combination fails at parse time with a clear usage error instead of a silent
no-op. Also note in `--output`'s help that it is not used with `--quick`
(mirroring how `--label` is already documented as quick-only).

## Related Issue

Fixes #98369

## Type of Change

## Changes Made

- `hermes_cli/subcommands/backup.py` — `-o/--output` and `-q/--quick` are now a
  mutually exclusive group; `--output` help notes it is not used with `--quick`.
- `tests/hermes_cli/test_backup_output_quick_exclusive.py` — the combo is
  rejected; valid combos still parse.

## How to Test

    pytest tests/hermes_cli/test_backup_output_quick_exclusive.py -q

4 pass. `backup --quick -o x` exits 2 with a usage error; `backup -o x`,
`backup --quick`, and `backup --quick -l label` still parse.

## Checklist

### DIFF
--- a/hermes_cli/subcommands/backup.py
+++ b/hermes_cli/subcommands/backup.py
@@ -21,12 +21,18 @@ def build_backup_parser(subparsers, *, cmd_backup: Callable) -> None:
         "skills, sessions, and data (excludes the hermes-agent codebase). "
         "Use --quick for a fast snapshot of just critical state files.",
     )
-    backup_parser.add_argument(
+    # -o/--output writes a full archive at that path; --quick instead stores a
+    # state snapshot into HERMES_HOME/state-snapshots/ and never writes an
+    # archive, so the two cannot combine. Passing both used to accept -o and
+    # silently never write the file (exit 0, nothing created) (#98369); make it
+    # a clear argparse error instead.
+    _backup_target = backup_parser.add_mutually_exclusive_group()
+    _backup_target.add_argument(
         "-o",
         "--output",
-        help="Output path for the zip file (default: ~/hermes-backup-<timestamp>.zip)",
+        help="Output path for the zip file (default: ~/hermes-backup-<timestamp>.zip); not used with --quick",
     )
-    backup_parser.add_argument(
+    _backup_target.add_argument(
         "-q",
         "--quick",
         action="store_true",
--- a/tests/hermes_cli/test_backup_output_quick_exclusive.py
+++ b/tests/hermes_cli/test_backup_output_quick_exclusive.py
@@ -1,0 +1,38 @@
+"""`hermes backup`: -o/--output and --quick are mutually exclusive (#98369).
+
+--quick stores a state snapshot into HERMES_HOME/state-snapshots/ and never
+writes an archive, so accepting -o with --quick silently dropped the requested
+path (exit 0, nothing created). The combination is now rejected at parse time.
+"""
+import argparse
+
+import pytest
+
+from hermes_cli.subcommands.backup import build_backup_parser
+
+
+def _parser() -> argparse.ArgumentParser:
+    p = argparse.ArgumentParser(prog="hermes")
+    sub = p.add_subparsers()
+    build_backup_parser(sub, cmd_backup=lambda args: None)
+    return p
+
+
+def test_quick_with_output_is_rejected():
+    with pytest.raises(SystemExit) as ei:
+        _parser().parse_args(["backup", "--quick", "-o", "/tmp/x.zip"])
+    assert ei.value.code == 2  # argparse usage error, not a silent exit 0
+
+
+@pytest.mark.parametrize(
+    "argv,expect",
+    [
+        (["backup", "-o", "/tmp/x.zip"], {"output": "/tmp/x.zip", "quick": False}),
+        (["backup", "--quick"], {"output": None, "quick": True}),
+        (["backup", "--quick", "-l", "nightly"], {"output": None, "quick": True, "label": "nightly"}),
+    ],
+)
+def test_valid_backup_arg_combos_still_parse(argv, expect):
+    ns = _parser().parse_args(argv)
+    for key, value in expect.items():
+        assert getattr(ns, key) == value


liyangbing:
<h3>Review</h3><p>This is the right user-facing boundary: <code>backup --quick -o path</code> cannot produce the requested archive, so accepting it and exiting 0 is worse than rejecting the combination. Moving the check into argparse also prevents the quick path from silently claiming that an output file was written.</p><p>Please keep the command contract precise:</p><p>1. Reject every spelling/ordering of the incompatible pair, including short options, long options, and repeated flags, before creating snapshots or mutating backup state.<br>2. Keep the error on stderr with a non-zero exit status and usage text that names both options and the valid alternatives.<br>3. Preserve the existing semantics of <code>backup -o path</code>, <code>backup --quick</code>, and quick-only labels; test that each still reaches the intended execution path rather than merely parsing.<br>4. Ensure a failed parse cannot leave a partial snapshot, archive, temp file, or misleading success marker behind.<br>5. Keep help text and shell completion/schema metadata consistent with the mutual exclusion, if those are generated or maintained separately.</p><h3>Regression matrix</h3><p>Please cover:<br>• <code>-q -o path</code>, <code>--quick --output path</code>, reversed ordering, and repeated flags<br>• missing output path, output path equal to an existing file, and an unwritable directory<br>• quick backup with and without <code>--label</code>, proving the snapshot still completes<br>• non-quick backup with output path, proving the archive is created at the requested path<br>• parse failure stdout/stderr, exit code, usage text, and absence of side effects<br>• Windows path quoting and paths containing spaces/non-ASCII characters</p><p>Direct evidence: <a href="https://github.com/NousResearch/hermes-agent/pull/98375">PR #98375</a> and its related <a href="https://github.com/NousResearch/hermes-agent/issues/98369">issue #98369</a>, which identify <code>hermes_cli/subcommands/backup.py</code> and the added parser tests.</p>

PR #36622, opened 2026-06-01
fix: avoid routing Feishu quoted replies into topics

mei959:
## Summary
- Stop Feishu quoted replies from being routed into topic/thread sessions by removing quoted-message root IDs from generic thread metadata.
- Keep Feishu thread routing based on explicit topic context rather than ordinary quote/reply context.
- Add regression coverage for Feishu quoted replies and Telegram thread fallback behavior.

## Test Plan
- python -m pytest tests/gateway/test_feishu.py tests/gateway/test_telegram_thread_fallback.py -q

## Notes
- Local working tree still has unrelated uncommitted package-lock changes in ui-tui/ and web/; they are not part of this PR.

PR #36233, opened 2026-06-01
fix(feishu): distinguish quote replies from topic sessions

SergioYin:
## Problem

In normal conversation, quoting a message is intended to inject or emphasize context within the current conversation.

Previously, Feishu reply metadata such as `root_id` could be promoted into the Hermes thread route. As a result, an ordinary quote reply could be routed into a separate topic-scoped session. From the user's perspective, quoting a message unexpectedly opened a fresh context instead of reinforcing the current one.

Quote context versus session routing

## What changed

This PR makes the distinction explicit in the active Feishu adapter at `plugins/platforms/feishu/adapter.py`:

- quote metadata remains reply context;
- an explicit `thread_id` selects a topic-scoped session;
- ordinary DM and group quote replies remain in the current session;
- the forum-only `root_id` compatibility fallback is preserved;
- quoted text and media remain available to the agent as context.

In short:

> Quoting adds context. It does not define a session boundary.

## Behavior

- **Ordinary quote reply:** continue the current DM/group session and attach the quoted message as context.
- **Explicit topic reply:** route through the Feishu `thread_id` and continue the topic-scoped session.
- **Legacy forum payload:** allow the scoped `root_id` fallback when an explicit `thread_id` is unavailable.

This avoids the previous behavior where quoting a message could silently move the user into a different Hermes session.

## Changes made

- `plugins/platforms/feishu/adapter.py`
  - Adds `_resolve_inbound_thread_id()` to centralize reply/topic routing semantics.
  - Separates the reply anchor from the session-routing identifier.
  - Preserves quoted text/media retrieval through the adapter's blocking-call executor.
- `tests/gateway/test_feishu.py`
  - Adds regressions for ordinary DM/group replies, explicit topics, quoted media, and forum fallback behavior.
- `tests/gateway/test_run_progress_topics.py`
  - Verifies ordinary Feishu quote replies do not receive topic progress metadata, while explicit topics retain it.
- `website/docs/user-guide/messaging/feishu.md`
  - Documents quote-reply versus topic-routing behavior.

## Related issue

Fixes #20548.

Related context: #20562, #29467, #30164, #20851.

## Type of change

- [x] Bug fix
- [x] Documentation update
- [x] Tests added or improved
- [ ] New feature
- [ ] Refactor without behavior change

## Validation

Relevant gateway test suite:

```text
377 passed
```

Additional checks:

```text
Ruff passed
Python compile checks passed
git diff --check passed
```

[...]

PR #98376, opened 2026-08-30
fix(feishu): preserve quoted context and native topic routing

wjykyhhh:
## Summary

Fix Feishu quoted-reply handling as one coherent behavior class:

- keep ordinary group messages and ordinary quoted replies in the main channel;
- keep native topic messages in their real Feishu thread and anchor replies to the current inbound message;
- inject quoted parent text and referenced image/file media into the current `MessageEvent`;
- dispatch quote-only messages whose body becomes empty after stripping the bot mention;
- retain the existing behavior that a pure bot mention with no quoted parent is ignored.

## User-visible failures

1. **A reply inside a native Feishu topic could anchor to the quoted/root message instead of the newest user message.** The bot response appeared under the wrong message in the thread.
2. **An ordinary group message or ordinary quoted reply could be misclassified as a topic.** `root_id` was used as a fallback for `thread_id`, so a main-channel response was sent into a reply thread.
3. **Quoted images/files lost their actual content.** Parent lookup returned text only; image parents could surface as `[Replying to: "None"]` or as a placeholder without the media resource, so vision-capable models could not inspect the quoted image.
4. **Quoting a message and sending only `@Bot` produced no response.** Self-mention stripping reduced the current body to an empty string, and the adapter returned before resolving the quoted parent.
5. **An unavailable/withdrawn quoted parent failed silently in the mention-only case.** There was no semantic payload left to dispatch and no explanation to the model.
6. **Native-topic media and generic error responses could lose the current-message anchor.** Text sends received `reply_to` directly, but metadata-only media/error paths had only `thread_id` and could fall back to `message.create(receive_id_type="thread_id")`.
7. **A top-level Feishu forum chat could still become a reply/topic.** The top-level guard covered `chat_type="group"` but not Feishu's `forum` source type.
8. **Frequently quoted cached media could become permanently unreadable after housekeeping.** The quoted-context LRU cached local paths indefinitely while hourly media cleanup could delete the underlying files.

## Root causes

- **Delivery scope and quote semantics were conflated.** Native `thread_id` determines topic scope; `root_id`, `parent_id`, and `upper_message_id` only describe the quoted-parent relationship.
- **Thread reply anchoring reused the quoted parent instead of the current inbound message.** A true topic response must reply to the newest user message while preserving the quoted parent separately as context.
- **Quoted-parent retrieval was text-only and stringified optional placeholders too early.** `str(None).strip()` produced the literal string `"None"`, and referenced media never entered the normal inbound media path.
- **The empty-text guard ran before quoted-context lookup.** Mention-only quoted replies were dropped before the parent could contribute text or media.
- **The shared thread metadata omitted Feishu's current reply anchor.** Non-text/error send paths therefore lacked enough information to call `message.reply(..., reply_in_thread=True)`.
- **Top-level suppression was keyed to one chat-type label instead of Feishu scope.** Non-DM forum/channel events without native `thread_id` fell through to `event.message_id`.
- **The quoted-context cache validated neither TTL nor local path existence.** A cache hit could outlive the downloaded resource it referenced.

## Fix

- Treat only a non-empty native Feishu `thread_id` as topic scope; never infer a topic from `root_id` or `parent_id`.
- Centralize reply-anchor selection so ordinary group sends remain top-level while native topics reply to the current inbound message.
- Add a structured quoted-message context carrying normalized text plus ordered media URLs/MIME types, backed by the existing bounded LRU cache.
- Download quoted image/file resources through the normal Feishu resource path and merge them into the current event.
- Preserve nulls before string conversion so missing placeholders cannot become the fake text `"None"`.
- Resolve quoted context before applying the empty-text guard. A mention-only quote dispatches when the parent contributes text/media; an unreadable parent gets `[Quoted message unavailable]`; an unquoted pure mention remains ignored.
- Keep batching compatibility strict across quoted parent and native thread identity.
- Carry `reply_to_message_id` in Feishu native-thread metadata so text, media, and error paths use the same current-message anchor.
- Treat every non-DM Feishu event without native `thread_id` as top-level, including group/forum/channel source types.
- Validate cached quoted-media paths on every hit; invalidate and refetch the parent context if housekeeping removed a file.

## Behavioral invariants

| Inbound Feishu event | Expected outbound behavior |
| --- | --- |
| Main-channel message | New top-level group message |
| Main-channel ordinary quote | Quoted parent is model context; response remains top-level |
| Native topic message | Response remains in that native `thread_id` and anchors to the current inbound message |
| Quoted image/file | Placeholder explains the relation; actual resource is attached through the normal media path |
| Quoted message + only `@Bot` | Dispatch using quoted text/media |
| Only `@Bot`, no quote | Ignore |
| Quoted parent unavailable | Dispatch with an explicit unavailable-parent marker |

## Tests

- Focused forum/thread-metadata/stale-cache/media-path checks: **9 passed**
- Reply-pointer gateway preprocessing (including empty current text): **3 passed**
- Feishu/routing regression set: **109 passed, 18 skipped**
- Extended Feishu set on the PR branch: **147 passed, 18 skipped**, with two existing approval-card failures
- The same two approval-card tests fail unchanged on current `upstream/main`, so this change introduces no new failure:
  - `test_rejects_approval_click_from_unauthorized_user`
  - `test_update_prompt_unauthorized_operator_returns_no_card`
- Ruff: passed
- `git diff --check`: passed
- Added-line security scan: no hardcoded secrets, shell injection, `eval`/`exec`, or unsafe pickle usage
- Independent read-only review: **passed after one fix/re-review cycle**; no remaining security or logic blockers

## Manual Feishu verification

Verified against a live Feishu workspace:

- main-channel text quote + mention-only body dispatched with the parent text;
- main-channel image quote + mention-only body attached one real native image and was readable by the vision model;
- native topic quote + mention-only body preserved its native `thread_id`;
- main-channel responses stayed top-level rather than creating accidental topics.

### DIFF
--- a/gateway/platforms/base.py
+++ b/gateway/platforms/base.py
@@ -151,6 +151,12 @@ def _thread_metadata_for_source(source, reply_to_message_id: str | None = None) 
         scope_id = getattr(source, "scope_id", None)
         if scope_id:
             metadata["slack_team_id"] = str(scope_id)
+    if (
+        _platform_name(getattr(source, "platform", None)) == "feishu"
+        and thread_id is not None
+        and reply_to_message_id is not None
+    ):
+        metadata["reply_to_message_id"] = str(reply_to_message_id)
     if not metadata:
         return None
     if _platform_name(getattr(source, "platform", None)) == "telegram" and getattr(source, "chat_type", None) == "dm":
@@ -207,10 +213,10 @@ def _reply_anchor_for_event(event) -> str | None:
             # the newest user message. A quoted parent remains only a recovery
             # fallback when a synthetic/resumed event has no current message ID.
             return getattr(event, "message_id", None) or getattr(event, "reply_to_message_id", None)
-        if getattr(source, "chat_type", None) == "group":
-            # A group-channel mention is a new top-level chat message unless the
-            # inbound event carries a real Feishu thread_id.  Passing message_id
-            # here would call message.reply and render the answer as a reply/topic.
+        if getattr(source, "chat_type", None) != "dm":
+            # A non-DM mention is a new top-level chat message unless the inbound
+            # event carries a real Feishu thread_id. Passing message_id here would
+            # call message.reply and render group/forum/channel answers as topics.
             return None
     return getattr(event, "message_id", None)
 
--- a/gateway/platforms/base.py
+++ b/gateway/platforms/base.py
@@ -201,8 +201,11 @@ def _reply_anchor_for_event(event) -> str | None:
         return getattr(event, "message_id", None) or getattr(event, "reply_to_message_id", None)
     if platform == "telegram" and thread_id:
         return None
-    if platform == "feishu" and thread_id and getattr(event, "reply_to_message_id", None):
-        return getattr(event, "reply_to_message_id", None)
+    if platform == "feishu" and thread_id:
+        # Keep the reply inside the topic via thread metadata, but anchor it to
+        # the newest user message. A quoted parent remains only a recovery
+        # fallback when a synthetic/resumed event has no current message ID.
+        return getattr(event, "message_id", None) or getattr(event, "reply_to_message_id", None)
     return getattr(event, "message_id", None)
 
 
--- a/gateway/platforms/base.py
+++ b/gateway/platforms/base.py
@@ -201,11 +201,17 @@ def _reply_anchor_for_event(event) -> str | None:
         return getattr(event, "message_id", None) or getattr(event, "reply_to_message_id", None)
     if platform == "telegram" and thread_id:
         return None
-    if platform == "feishu" and thread_id:
-        # Keep the reply inside the topic via thread metadata, but anchor it to
-        # the newest user message. A quoted parent remains only a recovery
-        # fallback when a synthetic/resumed event has no current message ID.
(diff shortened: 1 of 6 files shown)


alt-glitch:

Related: #36233 and #36622 already address the quote-reply/topic-routing family. This PR also adds quoted-media context and current-message reply anchoring, so it is an overlapping broader alternative rather than a duplicate.

liyangbing:
<h3>Review</h3><p>This PR separates Feishu native topic scope from quoted-parent context. The current inbound message must own the reply anchor; quoted content supplies context only.</p><p>Keep these boundaries explicit:<br>1. Use only a non-empty native thread_id for topic delivery; root_id and parent_id must not promote an ordinary group reply.<br>2. Preserve the current message_id through text, media, error, retry, and fallback sends.<br>3. Keep quoted text and media structured, with an explicit unavailable marker; never stringify missing values as None.</p><h3>Regression matrix</h3><p>Cover main-channel text and ordinary quotes, native topics, quote-only Bot messages, unavailable parents, all outbound paths, stale-media cleanup races, concurrent chats, retry after restart, and pure unquoted Bot messages remaining ignored.</p><p>Direct evidence: PR 98376 at https://github.com/NousResearch/hermes-agent/pull/98376, including its Feishu adapter and routing changes.</p>

PR #6622, opened 2026-04-09
fix: error context preservation, WAL checkpoint, hook timeout

aaronlab:
## Summary

This PR addresses three reliability gaps found during iteration #4 of a deep code audit:

- **Preserve original error context in call_llm retry chain** (`agent/auxiliary_client.py`): When the `max_tokens` retry also fails with a payment error, the original error was silently overwritten via `first_err = retry_err` (line 2089), losing the initial diagnostic context. Now uses `retry_err.__cause__ = first_err` for proper Python exception chaining, preserving the full error sequence for debugging.

- **Guard payment fallback provider call** (`agent/auxiliary_client.py`): The fallback API call (line 2105) had no try/except. If the fallback provider also failed, there was zero logging and no indication that a fallback was even attempted. Added error handling with a warning log message identifying which fallback provider failed.

- **Add WAL checkpoint to holographic memory store** (`plugins/memory/holographic/store.py`): WAL mode was enabled (line 130) but no checkpoint mechanism existed anywhere in the class, unlike `hermes_state.py` which has proper `_try_wal_checkpoint()`. This causes unbounded WAL file growth over long-running sessions. Added periodic PASSIVE checkpoint every 50 writes and a final checkpoint on `close()`, following the established pattern.

- **Add timeout protection to plugin hook invocation** (`hermes_cli/plugins.py`): Hook callbacks had good exception isolation (try/except per callback) but no timeout protection. A misbehaving plugin could block the agent loop indefinitely with a blocking call. Added a 30-second timeout using `ThreadPoolExecutor`, with warning logging on timeout.

## Files Changed

| File | Change |
|------|--------|
| `agent/auxiliary_client.py` | Exception chaining for error context + fallback error handling |
| `plugins/memory/holographic/store.py` | WAL checkpoint method + periodic + on-close checkpoint |
| `hermes_cli/plugins.py` | 30s timeout for hook callbacks via ThreadPoolExecutor |

## Test plan

- [ ] Simulate primary provider max_tokens error → payment retry → verify original error in `__cause__`
- [ ] Simulate fallback provider failure → verify warning log is emitted
- [ ] Insert 100+ facts into holographic store → verify WAL file stays bounded
- [ ] Create a plugin with `time.sleep(60)` in pre_llm_call hook → verify 30s timeout warning
- [ ] Run existing test suite: `pytest tests/`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #98385, opened 2026-08-30
fix(plugins): don't drop healthy concurrent fail-open hook invocations

### LINKED ISSUE #98382: Concurrent observer-hook invocations are dropped as if a callback had timed out
Bounded hooks use one `(hook_name, id(callback))` running bit for two different states: a callback that exceeded `plugins.hook_callback_timeout`, and a healthy callback still processing another invocation. In `PluginManager.invoke_hook`, lines around the `running = callback_key in self._hook_running_callbacks` check skip either state with the same warning. For fail-open observer hooks such as `post_tool_call`, a second concurrent event is therefore silently discarded even when the first callback completes well within its timeout.

This is reproducible without a timeout: register a `post_tool_call` callback that takes ~0.6s, invoke the hook from two threads 0.3s apart, and the callback runs once. The second event logs `skipped after previous timeout or while still running`. On a profile with overlapping cron sessions this dropped 376 callback events in two days while the log recorded zero actual hook timeouts. The affected callbacks were evidence/audit recorders, so this loses telemetry rather than merely delaying it.

The timeout guarantee should remain: a genuinely timed-out worker must not spawn unbounded duplicates, and `pre_tool_call` must continue to fail closed. But a healthy in-flight invocation needs a different policy, likely per-invocation workers or a bounded per-callback queue for observer hooks. A regression test should start two healthy concurrent invocations and assert both payloads are processed, alongside the existing `test_hung_callback_suppresses_repeat_fires` timeout case.

chelsealong:
Fixes #98382

## Problem

`PluginManager.invoke_hook` used a single `running` flag
(`callback_key in self._hook_running_callbacks`) to mean two different
things: "this callback's worker exceeded `plugins.hook_callback_timeout`
and is still running" and "a healthy invocation of this callback from a
moment ago hasn't returned yet." Both states hit the same skip branch:

```python
if (
    suppressed_until is not None and suppressed_until > now
) or running:
    logger.warning(...)
    if fail_closed:
        results.append(_pre_tool_call_timeout_block())
    continue
```

For fail-open observer hooks (everything except `pre_tool_call`), this
meant a second concurrent event was silently dropped even when the first
callback was well within its timeout — e.g. two `post_tool_call` events
0.3s apart, callback duration ~0.6s, default timeout 30s: no timeout ever
occurred, yet the second invocation was skipped and logged as "skipped
after previous timeout or while still running." Reported as 376 dropped
callback events in two days on a profile with overlapping cron sessions,
with the affected callbacks being evidence/audit recorders — so this
silently loses telemetry.

## Fix

Split the two states apart in `hermes_cli/plugins.py::PluginManager.invoke_hook`:

- `timed_out` — a *confirmed* timeout (from `_hook_timeout_suppressed_until`,
  set only when a caller's `done.wait(timeout)` actually elapsed) — always
  skips, for every hook. This is what keeps a genuinely hung worker from
  spawning unbounded duplicates.
- `running` alone (worker still executing, no confirmed timeout) — only
  skips for `pre_tool_call`, which must keep failing closed on any overlap.
  For other bounded hooks, a second invocation now starts its own
  concurrent worker instead of being dropped.

`pre_tool_call`'s fail-closed behavior is unchanged (still covered by
`test_pre_tool_call_timeout_fail_closed`), and the existing hung-callback
suppression test (`test_hung_callback_suppresses_repeat_fires`) still
passes because it exercises the confirmed-timeout path, not mere overlap.

## Test

Added `test_concurrent_healthy_invocations_both_processed` in
`tests/hermes_cli/test_plugins.py`: registers a `post_tool_call` callback
that sleeps 0.3s, invokes the hook from two threads 0.15s apart with a
5s timeout budget, and asserts both invocations actually ran and both
results came back.

Confirmed it fails without the fix (`git checkout HEAD~1 -- hermes_cli/plugins.py`):

```
FAILED tests/hermes_cli/test_plugins.py::TestForceReloadSymmetry::test_concurrent_healthy_invocations_both_processed
AssertionError: assert ['first'] == ['first', 'second']
...
WARNING  hermes_cli.plugins:plugins.py:5625 Hook 'post_tool_call' callback slow skipped after previous timeout or while still running
```

With the fix, full suite passes via the mandated hermetic runner:

```
$ scripts/run_tests.sh tests/hermes_cli/test_plugins.py
[100.0% |    75/~75 | ✓75 | ✗ 0] ✓ tests/hermes_cli/test_plugins.py (75✓, 4.3s)
=== Summary: 1 files, 75 tests passed, 0 failed (100% complete) in 4.3s (8 workers) ===
```

`ruff check hermes_cli/plugins.py tests/hermes_cli/test_plugins.py` — all checks passed.

## AI assistance disclosure

This change was authored by an AI coding agent (Claude), including the
diagnosis, fix, and regression test, with the test-failure-without-fix
verification shown above run by the same agent.

### DIFF
--- a/hermes_cli/config_defaults.py
+++ b/hermes_cli/config_defaults.py
@@ -2603,6 +2603,10 @@ DEFAULT_CONFIG = {
         # callback. Shell hooks keep their own per-entry ``timeout``. Set to 0
         # to disable the cap (sync call on the agent thread). Capped at 600.
         "hook_callback_timeout": 30,
+        # Max concurrent in-flight workers per (hook, callback) for fail-open
+        # hooks before further invocations are dropped and counted as
+        # overflow rather than run unbounded. Capped at 64.
+        "hook_callback_max_concurrency": 4,
     },
 
     # Shell-script hooks — declarative bridge that invokes shell scripts
--- a/hermes_cli/plugins.py
+++ b/hermes_cli/plugins.py
@@ -3669,6 +3669,47 @@ class PluginContext:
 _HOOK_CALLBACK_TIMEOUT_SECS = 30.0
 _MAX_HOOK_CALLBACK_TIMEOUT_SECS = 600.0
 
+# Cap on concurrent in-flight workers per (hook, callback) for fail-open
+# hooks, so a sustained event burst cannot fan out unbounded workers once a
+# callback's duration exceeds its arrival rate. Overridden by
+# ``plugins.hook_callback_max_concurrency``. Does not apply to fail-closed
+# hooks (``pre_tool_call``), which already skip on any overlap.
+_HOOK_CALLBACK_MAX_CONCURRENCY = 4
+_MAX_HOOK_CALLBACK_MAX_CONCURRENCY = 64
+
+
+def _resolve_hook_callback_max_concurrency() -> int:
+    """Return the effective per-callback concurrency cap for fail-open hooks.
+
+    Reads ``plugins.hook_callback_max_concurrency`` via the cached readonly
+    config loader. Falls back to ``_HOOK_CALLBACK_MAX_CONCURRENCY``. Values
+    ``< 1`` are clamped to 1; values above
+    ``_MAX_HOOK_CALLBACK_MAX_CONCURRENCY`` are clamped down.
+    """
+    limit = _HOOK_CALLBACK_MAX_CONCURRENCY
+    try:
+        from hermes_cli.config import load_config_readonly
+
+        plugins_cfg = (load_config_readonly() or {}).get("plugins")
+        if isinstance(plugins_cfg, dict) and "hook_callback_max_concurrency" in plugins_cfg:
+            raw = plugins_cfg.get("hook_callback_max_concurrency")
+            if raw is not None:
+                limit = int(raw)
+    except (TypeError, ValueError):
+        logger.warning(
+            "plugins.hook_callback_max_concurrency is not an int; using default %d",
+            _HOOK_CALLBACK_MAX_CONCURRENCY,
+        )
+        limit = _HOOK_CALLBACK_MAX_CONCURRENCY
+    except Exception:
+        limit = _HOOK_CALLBACK_MAX_CONCURRENCY
+
+    if limit < 1:
+        return 1
+    if limit > _MAX_HOOK_CALLBACK_MAX_CONCURRENCY:
+        return _MAX_HOOK_CALLBACK_MAX_CONCURRENCY
+    return limit
+
 
 def _resolve_hook_callback_timeout() -> float:
     """Return the effective hook-callback timeout in seconds.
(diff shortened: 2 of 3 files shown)


liyangbing:
<h3>Review</h3><p>The fix correctly separates a confirmed timeout from a healthy overlap, and the mutation check is strong evidence for the original regression. One follow-up boundary remains: allowing every bounded fail-open hook to start another worker removes silent loss, but without a per-callback concurrency/queue limit it can turn a burst of events into unbounded worker growth when callback duration exceeds arrival rate.</p><p>Please make the post-fix policy explicit:</p><p>1. Define a bounded maximum of concurrent workers or queued events per callback/hook, with an observable overflow result and counters.<br>2. Keep the confirmed-timeout suppression separate from ordinary backpressure; queue saturation must not be mislabeled as a timeout.<br>3. Preserve event ordering where an observer depends on it, or document that callbacks may complete out of order and carry an event sequence for reconciliation.<br>4. Define cancellation and shutdown: admitted callbacks must either drain within a bound or be reported as dropped, and a late worker must not mutate unloaded plugin state.<br>5. Keep <code>pre_tool_call</code> fail-closed on overlap and confirmed timeout, while proving observer exceptions/cancellation do not change that policy.</p><h3>Regression matrix</h3><p>Please cover:<br>• two healthy concurrent invocations: both run and return<br>• sustained event burst with slow callback: bounded workers/queue and explicit overflow evidence<br>• confirmed timeout followed by repeated events: no worker fan-out<br>• timeout watchdog racing with normal completion<br>• out-of-order completion with event IDs and ordering policy<br>• plugin reload/unload and process shutdown while workers or queued events remain<br>• concurrent <code>pre_tool_call</code> overlap, callback exception, cancellation, and fail-closed result</p><p>Direct evidence: <a href="https://github.com/NousResearch/hermes-agent/pull/98385">PR #98385</a>, including the <code>PluginManager.invoke_hook</code> change and the mutation-verified healthy-concurrency test.</p>

chelsealong:
Addressed the concrete, boundable part of the follow-up: added `plugins.hook_callback_max_concurrency` (default 4) to cap in-flight workers per `(hook, callback)` for fail-open hooks, so a sustained burst can no longer fan out unbounded workers once callback duration exceeds arrival rate. Overflow beyond the cap is dropped and counted in a separate `_hook_overflow_counts` dict, keeping it distinct from a confirmed timeout (item 2). Docstring now documents that concurrent workers for the same callback may complete out of order (item 3's documented alternative, rather than adding event-sequence plumbing). `pre_tool_call` is untouched — it already fails closed on any overlap regardless of this cap, so item 5's policy is unchanged (existing `test_pre_tool_call_timeout_fail_closed` still covers it). Added `test_burst_beyond_max_concurrency_is_bounded_and_counted`, which fails on `HEAD~1` with `AttributeError: ... no attribute '_resolve_hook_callback_max_concurrency'` and passes with the fix; full `tests/hermes_cli/test_plugins.py` is green (76/76) via `scripts/run_tests.sh`, and `ruff check` is clean.

Left out of scope for this PR: item 4 (drain-on-shutdown / cancellation reporting). Implementing a bounded drain would require joining the abandoned worker threads this file deliberately never joins, to avoid reintroducing the #6622 hang — that's a real conflict with an existing invariant here, not something I want to guess a resolution for without maintainer input, so I didn't add code for it.

kshitijk4poor:
Thanks @chelsealong — the symptom this fixes (healthy concurrent invocations of the same fail-open hook being dropped by the per-callback running gate) is now resolved on main by #111177 (`73f808e47f`, salvage of #110470 by @deadczarvc): the gate is keyed on the call identity (`tool_call_id`/`turn_id`), a timed-out worker is tracked as abandoned per callback so a hung hook still leaks exactly one thread, and `transform_terminal_output`/`transform_llm_output` carry an identity too. Bound by `test_concurrent_same_tool_calls_with_distinct_ids_both_run`. Closing as superseded; you submitted this before that fix existed, so credit for the report and shape stands.

PR #98388, opened 2026-08-30
fix(desktop): fail closed when Bot Chat lookup returns zero rows

### LINKED ISSUE #98383: [Bug] Bot Mode mints a new canonical Bot Chat + re-fires kickoff on click despite an intact registry row (post-#95326)
## Summary

On a fully current checkout, clicking a bot in the Bots pane **mints a new canonical Bot Chat and re-fires the kickoff intro** instead of resolving the existing one — even though the registry row is intact and every resolver returns it correctly when called directly.

This is the symptom class of #92473 / #90705 / #93056 / #92687, but **none of those root causes apply here**: the row is not archived, not retitled, not pointer-pinned, and the `kickoff = false` click-path guard (the #95326 fix referenced in #94726) **is present in the running bundle**. Six sessions were minted in ~90 minutes, each stamped with a user-attributed `Hey, tell me about yourself!`, orphaning ~39 messages of in-progress work into a session the UI no longer surfaces.

Filing because #94726 lists the click-path re-fire as killed by #95326, and this reproduces downstream of that fix.

## Environment

- Hermes Agent @ `4209d371a` (`origin/main`, behind only by `26350357d`, a telegram test change)
- Desktop bundle `dist/assets/index-Dh-pjnId.js`, sha-identical in `dist/`, `app.asar.unpacked/`, and **inside `app.asar`** (verified — the app is not loading a stale build)
- Gateway process started 09:37, after the last source mtime (09:36) — no version skew
- macOS 26.6.2, profile `henry`, local (non-remote) mode

## What happened

Six sessions on one profile in ~90 minutes. Every one begins with the identical 28-character user message:

```
id                      title                 msgs  hidden
20260830_130804_33f034  Bot Chat              213   1      <- the real canonical row
20260830_131806_a10f14  Friendly greeting     2     1
20260830_131916_994b92  Friendly greeting #2  2     1
20260830_141358_286dd8  Friendly greeting #3  4     1
20260830_143139_bdb8af  Friendly greeting #4  39    1      <- orphaned real work
20260830_143742_10e15e  Friendly greeting #5  14    1
```

```sql
sqlite> select session_id, content from messages where role='user' and length(content)=28;
20260830_130804_33f034|Hey, tell me about yourself!
20260830_131806_a10f14|Hey, tell me about yourself!
20260830_131916_994b92|Hey, tell me about yourself!
20260830_141358_286dd8|Hey, tell me about yourself!
20260830_143139_bdb8af|Hey, tell me about yourself!
20260830_143742_10e15e|Hey, tell me about yourself!
```

That string is `kickoffText()` — so each click reached `createCanonicalChat(..., {kickoff: true})`, which per the source comment is New-Agent-only.

Each new session also became the profile's `last_session`, so the roster preview followed the newest orphan and the 213-message Bot Chat became unreachable by clicking. From the user's side this reads as "all history is deleted when I click the main session and come back."

## Why the known root causes are excluded

The registry row is healthy — the exact server-side resolvers, run against the live DB:

```
BY_TITLE: {'id': '20260830_130804_33f034', 'source': 'tui', 'archived': 0,
           'hidden': 1, 'title': 'Bot Chat', 'message_count': 213}
TIP:      20260830_130804_33f034
```

- **Not archived** (`archived=0`) → excludes #93056 / #92687
- **Not retitled**, `title_source='user'`, `idx_sessions_title_unique` present → excludes #92473
- **`ui_meta['hermes-bots'].chat` is `null`**, which is correct on current main (pointer identity was removed; legacy keys ignored) → excludes #90705
- **`hidden=1`** is expected, and `session.list` title-lookup resolves hidden rows

`_db_for_profile('henry')` returns the right DB, and the title lookup through it succeeds:

```
db for 'henry': /Users/davidsperling/.hermes/profiles/henry/state.db  owns=False
title lookup:  {'id': '20260830_130804_33f034', 'archived': 0, 'hidden': 1, 'message_count': 213}
```

The shipped bundle sends the correct registry query:

```js
jv(t,`session.list`,{profile:Av(r,n),title:Ny,limit:200,include_hidden:!0})
```

with ``Ny=`Bot Chat` ``, and its **only** `kickoff:!0` call site is the New Agent creation path — the click path passes no kickoff. So the guard is compiled in and still the intro fired.

## The gap

`findExistingCanonicalChat()` is documented to fail **closed**:

> FAIL CLOSED. A failed registry lookup MUST NOT read as "no Bot Chat exists" — that is the one remaining way to fork a bot's forever chat. The failure lives exactly in the post-update window: the desktop restarts every profile backend, the first bot click races the warm-up, the lookup RPC fails transiently...

Observed behaviour is consistent with that documented race — the desktop had just restarted the profile backends — but the mint happened **six times over 90 minutes**, long past any warm-up window, and the throw did not prevent a kickoff-bearing session from being created. So either the lookup is returning a successful-but-empty result (which is indistinguishable from "no Bot Chat" and bypasses the fail-closed throw entirely), or a second path reaches `createCanonicalChat` with kickoff set.

The empty-result case looks like the more likely of the two: the `catch` only fires on a **thrown** RPC error, while `rows.find(...) || null` maps an empty `sessions` array to the same `null` that means "mint one". A transient result-shape failure (backend mid-restart returning an empty list rather than an error) would therefore be read as authoritative absence.

**Suggested direction:** treat "registry lookup returned zero rows" as distinct from "registry lookup says this bot has no chat" on a profile whose `canonical_session` was non-null in the last known roster snapshot — i.e. require positive evidence of absence before minting, not merely the lack of positive evidence of presence.

## Reproduction

Not reliably reproducible on demand — every resolver invoked directly succeeds, which is precisely why the failure is invisible to the fail-closed guard. It reproduced 6/6 times on this install during normal clicking between the main Hermes session and a bot, immediately after a desktop restart.

If it would help, I can add instrumentation on `findExistingCanonicalChat` (log the raw `session.list` response before the `.find()`) and report the actual payload the next time it mints. Happy to do that if you'd like the empty-vs-throw question settled with live data rather than inference.

## Impact

Silent data-orphaning of active work. Nothing is deleted — everything is recoverable from `state.db` — but the UI offers no path back, and the natural reading is total history loss. The orphaned session here held 39 messages of an in-progress infrastructure audit.

chelsealong:
Fixes #98383.

## What changed and why

`findExistingCanonicalChat()` (`apps/desktop/src/plugins/hermes-bots/canonical-chat.ts`)
already fails closed on a **thrown** `session.list` error — it never mints a
replacement Bot Chat in that case. But a profile backend mid-restart can
answer `session.list` **successfully with an empty `sessions` array** instead
of throwing, and the old code:

```ts
const rows = res?.sessions ?? []
return rows.find(row => isCanonicalBotChatHistory(row)) || null
```

maps that empty result to the exact same `null` that means "this bot has
never had a chat yet." The click path (`openBotCanonicalChat` →
`createCanonicalChat`) reads `null` as "no row → create it," so it minted a
brand-new "Bot Chat" session and re-submitted the kickoff intro — even though
the real, hidden, 213-message canonical row was intact the whole time. The
issue reporter saw this happen 6 times in ~90 minutes after a desktop
restart, orphaning 39 messages of in-progress work into an unreachable
session each time.

The fix distinguishes "confirmed absent" from "unconfirmed absent": the
roster's own `canonical_session` field (already fetched server-side via
`profiles.list`, and already read elsewhere for preview/activity/the
`/new`→`/compact` guard — see AGENTS.md's Bot Mode section) is the last
positive confirmation that this profile has a Bot Chat. When that's present
but the lookup comes back with zero rows, treat it as an unconfirmed/failed
lookup and throw the same fail-closed error a thrown RPC error already
produces, instead of returning `null` and letting the caller mint. Both
existing call sites (`plugin.tsx`, `roster-actions.ts`) already catch this
error and show a "try again" toast — no new UI plumbing needed.

This does not introduce any stored session-id pointer (which AGENTS.md's Bot
Mode section explicitly forbids reintroducing) — it only reads the
already-server-resolved `canonical_session` that ships on every roster row.

## How to test

Reproduction (from the issue): click a bot row while its profile backend is
mid-restart such that `session.list` returns `{ sessions: [] }` instead of
throwing, on a profile whose roster row already carries a `canonical_session`
— previously this minted a new "Bot Chat" and fired the kickoff intro;
now it throws and the existing "try again" toast fires instead.

Added a regression test,
`apps/desktop/src/plugins/hermes-bots/canonical-chat-registry.test.ts`
→ *"refuses to mint on an empty lookup when the roster already confirmed a
canonical chat"*, which fails without the fix (falls through to
`session.create`, tripping the test's "must not create" guard) and passes
with it.

```
$ npx vitest run src/plugins/hermes-bots/canonical-chat-registry.test.ts
 Test Files  1 passed (1)
      Tests  9 passed (9)

$ npx vitest run src/plugins/hermes-bots
 Test Files  58 passed (58)
      Tests  560 passed (560)

$ npx eslint src/plugins/hermes-bots/canonical-chat.ts src/plugins/hermes-bots/canonical-chat-registry.test.ts
(clean)

$ npx tsc -p . --noEmit
(clean)
```

Confirmed the new test fails on the pre-fix source (`git checkout HEAD~1 --
apps/desktop/src/plugins/hermes-bots/canonical-chat.ts`):

```
AssertionError: expected [Function] to throw error matching /Bot Chat registry/
but got 'must not create: an empty result is not confirmed absence'
```

## Platforms tested

Linux (test suite only — no Desktop UI available in this environment).

## AI assistance disclosure

This patch was prepared with AI assistance (an autonomous coding agent), with
the diff, tests, and test output verified by running them locally as shown
above.

### DIFF
--- a/apps/desktop/src/plugins/hermes-bots/canonical-chat-registry.test.ts
+++ b/apps/desktop/src/plugins/hermes-bots/canonical-chat-registry.test.ts
@@ -280,4 +280,29 @@ describe('a failed lookup fails CLOSED — never "no chat exists"', () => {
     await expect(createCanonicalChat('ops')).rejects.toThrow(/Bot Chat registry/)
     expect(calls.some(call => call.method === 'session.create')).toBe(false)
   })
+
+  // #98383: a profile backend mid-restart can answer `session.list`
+  // SUCCESSFULLY with an empty list instead of throwing. `rows.find(...) ||
+  // null` used to read that identically to "this bot never had a chat",
+  // which minted a replacement and re-fired the kickoff on every click.
+  it('refuses to mint on an empty lookup when the roster already confirmed a canonical chat', async () => {
+    const calls = respondWith(method => {
+      if (method === 'session.list') {
+        return { sessions: [] }
+      }
+
+      if (method === 'session.create') {
+        throw new Error('must not create: an empty result is not confirmed absence')
+      }
+
+      return {}
+    })
+
+    const bot = { canonical_session: { id: 'forever-chat' }, name: 'ops' } as RosterRow
+    const { openBotCanonicalChat } = await loadModule()
+
+    await expect(openBotCanonicalChat(bot)).rejects.toThrow(/Bot Chat registry/)
+    expect(calls.some(call => call.method === 'session.create')).toBe(false)
+    expect(hostMock.openSession).not.toHaveBeenCalled()
+  })
 })
--- a/apps/desktop/src/plugins/hermes-bots/canonical-chat.ts
+++ b/apps/desktop/src/plugins/hermes-bots/canonical-chat.ts
@@ -211,8 +211,25 @@ async function findExistingCanonicalChat(owner: RosterRow | string): Promise<Can
   }
 
   const rows = res?.sessions ?? []
+  const match = rows.find(row => isCanonicalBotChatHistory(row))
 
-  return rows.find(row => isCanonicalBotChatHistory(row)) || null
+  if (match) {
+    return match
+  }
+
+  // A zero-row result is NOT the same as a thrown error, but it is just as
+  // capable of forking the forever chat: a profile backend mid-restart can
+  // answer `session.list` successfully with an empty list rather than
+  // failing it, and `|| null` used to read that identically to "this bot
+  // never had a chat" (#98383). The roster's own `canonical_session` is the
+  // last positive confirmation this profile HAD one — when that exists,
+  // an empty lookup is unconfirmed absence, not confirmed absence, so fail
+  // closed the same way a thrown RPC error already does instead of minting.
+  if (bot?.canonical_session?.id) {
+    throw new Error(`Could not confirm ${name}'s Bot Chat registry — not starting a new chat`)
+  }
+
+  return null
 }
 
 interface CreateCanonicalChatOptions {


Enough1122:

**Overall:** Fail-closed fix for Bot Chat lookup returning zero rows.

**What it does**
- `apps/desktop/src/plugins/hermes-bots/canonical-chat.ts:findExistingCanonicalChat` now distinguishes empty `sessions: []` from null: if no match and `bot?.canonical_session?.id` exists (roster already confirmed a forever chat), throws registry error instead of returning null and minting a replacement — mirrors existing thrown-error fail-closed path.
- Test adds case where roster has canonical_session but lookup returns empty array, asserting no `session.create` and no host open.

**Non-blocking notes**
- Uses `throw new Error("Could not confirm ... registry")` — matches existing thrown-error handling upstream; ensure error message is user-visible actionable (it propagates to caller catch that shows retry, good).

No functional issues found.

teknium1:
Thanks — this landed on main via #101210 (2599793271) with your commit/credit preserved. Clean cherry-pick. Closing the original in favor of the merged salvage.

PR #98389, opened 2026-08-30
fix(desktop-update): fall back to .venv in the posix hand-off (#98384)

### LINKED ISSUE #98384: [Bug]: macOS desktop update aborts "venv/bin/hermes is missing" on shared-venv layout (posix.sh hardcodes <root>/venv, no .venv fallback)
## Bug Description

On macOS git-installs that use the **shared-venv layout** (the actual venv lives at `~/.hermes/venvs/hermes`, and the checkout contains only a `.venv -> ~/.hermes/venvs/hermes` symlink — the layout `hermes doctor` itself reports as healthy), the desktop update hand-off always aborts at `scripts/desktop-update/posix.sh:564` because that gate hardcodes `$INSTALL_ROOT/venv/bin/hermes` (no dot-prefixed fallback):

```bash
HERMES_BIN="$INSTALL_ROOT/venv/bin/hermes"
[ -x "$HERMES_BIN" ] || { FINAL_CODE=3 FINAL_MSG="Update aborted: $HERMES_BIN is missing. The install needs repair (run the Hermes installer or hermes doctor)."; log "$FINAL_MSG"; exit 3; }
```

`posix.sh:237` has the same hardcoded no-dot path (`py="${INSTALL_ROOT:+$INSTALL_ROOT/venv/bin/python3}"`).

## Inconsistency with the rest of the codebase

The `.venv` fallback convention is already honored elsewhere, which makes this gate the odd one out:

- `hermes doctor` accepts both: it prints `✓ Venv entry point exists (.venv/bin/hermes)` on this layout (`hermes_cli/doctor.py` mentions `venv/bin/ or .venv/bin/`).
- `hermes_cli/managed_uv.py` (`_default_live_venv`) probes `<root>/venv` first and **falls back to `<root>/.venv`** explicitly.
- `hermes_cli/_install_repair.py` documents both conventions.

## Steps to Reproduce

1. Install via git with the shared-venv/uv-default layout: checkout at `~/.hermes/hermes-agent`, venv at `~/.hermes/venvs/hermes`, `.venv` symlink in the checkout, no `venv/` directory.
2. Trigger a desktop update (Check for updates / update button).
3. Update aborts before any git operation with exit code 3 and the message above.

## Expected Behavior

`posix.sh` should resolve the venv entry point the same way the rest of the updater does — probe `venv/bin/hermes`, fall back to `.venv/bin/hermes` (mirroring `_default_live_venv()` / doctor) — before declaring "the install needs repair".

## Additional findings

- Workaround: `ln -s ~/.hermes/venvs/hermes ~/.hermes/hermes-agent/venv` unblocks the gate (the update then completes successfully, proving everything else works).
- However, the managed-venv rotation logic (which tracks `venv/`, `venv.old/`, `venv.stale.runtime-*` per the updated `.gitignore` and `managed_uv.py` rename dance) can remove the hand-made symlink during a subsequent successful update, so the failure loop returns. A `.venv` fallback in `posix.sh` is the durable fix; the symlink workaround is only temporary.

## Environment

- OS: macOS (Apple Silicon, macOS 26.x)
- Hermes Agent v0.20.6 (2026.8.27), install method: git
- Layout: shared venv (`~/.hermes/venvs/hermes`) + `.venv` symlink in checkout (doctor-verified healthy)

chelsealong:
## Summary

Fixes #98384.

On macOS/Linux git-installs that use the shared-venv layout (the actual venv
lives at `~/.hermes/venvs/hermes`, and the checkout only has a `.venv ->
~/.hermes/venvs/hermes` symlink — a layout `hermes doctor` itself reports as
healthy), the desktop update hand-off in `scripts/desktop-update/posix.sh`
always aborted with:

> Update aborted: <root>/venv/bin/hermes is missing. The install needs
> repair (run the Hermes installer or hermes doctor).

`posix.sh` had two gates that only ever probed `venv/bin/`, with no
`.venv/bin/` fallback:

- `HERMES_BIN="$INSTALL_ROOT/venv/bin/hermes"` (the update gate, line ~564)
- `py="${INSTALL_ROOT:+$INSTALL_ROOT/venv/bin/python3}"` (the shim-UI
  interpreter, `start_ui()`, line ~237)

This is inconsistent with the rest of the codebase: `hermes doctor` and
`hermes_cli/managed_uv.py`'s `_default_live_venv()` both already probe `venv`
first and fall back to `.venv`. This PR makes `posix.sh` follow the same
precedence at both gates instead of telling a doctor-verified-healthy install
that it needs repair.

## Changes

- `scripts/desktop-update/posix.sh`: fall back to `.venv/bin/hermes` /
  `.venv/bin/python3` when `venv/bin/...` isn't executable, mirroring
  `_default_live_venv()`.
- `tests/test_desktop_update_shim_progress.py`: parameterize the existing
  end-to-end hand-off harness (`_run_handoff`) by venv directory, and add
  `test_dot_venv_layout_is_not_told_it_needs_repair`, which runs the real
  `posix.sh` against a `.venv`-only install root and asserts the update
  actually runs (rather than being rejected before `hermes` is ever invoked).

## Test plan

New test, run directly with `pytest` (no managed venv available in this
sandbox, so `scripts/run_tests.sh` itself wasn't runnable, but the test uses
the same real-subprocess harness as the existing suite in this file):

Before the fix (`git checkout HEAD~1 -- scripts/desktop-update/posix.sh`):
```
tests/test_desktop_update_shim_progress.py::test_dot_venv_layout_is_not_told_it_needs_repair FAILED
    assert len(stages) == 1
E   assert 0 == 1
E    +  where 0 = len([])
```
(0 stages seen — the gate rejected the `.venv`-only install before `hermes`
was ever invoked, i.e. the reported bug.)

After the fix:
```
tests/test_desktop_update_shim_progress.py::test_dot_venv_layout_is_not_told_it_needs_repair PASSED
======================= 6 passed, 3 errors in 5.10s ==========================
```

The 3 errors are pre-existing and unrelated: they come from
`tests/conftest.py`'s live-system `os.kill` guard firing during the
`progress` fixture's `proc.kill()` teardown for the three tests that use
that fixture. Confirmed identical on unmodified `HEAD` (5 passed, same 3
errors) before this change touched anything.

Also ran `shellcheck --severity=error scripts/desktop-update/posix.sh`: one
pre-existing SC2071 finding at an unrelated line (539), untouched by this
diff.

## AI assistance disclosure

This PR was prepared with AI assistance (an autonomous coding agent), with
the fix, test, and verification steps described above reviewed before
submission.

### DIFF
--- a/scripts/desktop-update/posix.sh
+++ b/scripts/desktop-update/posix.sh
@@ -235,6 +235,7 @@ start_ui() {
   [ "$NO_UI" -eq 1 ] && return
   local html="$SCRIPT_DIR/ui.html" py browser port="" i
   py="${INSTALL_ROOT:+$INSTALL_ROOT/venv/bin/python3}"
+  [ -x "${py:-/nonexistent}" ] || py="${INSTALL_ROOT:+$INSTALL_ROOT/.venv/bin/python3}"
   [ -x "${py:-/nonexistent}" ] || py="$(command -v python3 2>/dev/null)"
   browser="$(find_browser)"
   if [ -n "$browser" ] && [ -n "$py" ] && ! default_browser_is_chromium "$py"; then
@@ -562,6 +563,10 @@ sleep 1
 start_ui
 
 HERMES_BIN="$INSTALL_ROOT/venv/bin/hermes"
+# uv-default and shared-venv checkouts (doctor-verified healthy) use .venv
+# instead; mirror managed_uv.py's _default_live_venv() precedence rather than
+# telling a healthy install it needs repair.
+[ -x "$HERMES_BIN" ] || HERMES_BIN="$INSTALL_ROOT/.venv/bin/hermes"
 [ -x "$HERMES_BIN" ] || { FINAL_CODE=3 FINAL_MSG="Update aborted: $HERMES_BIN is missing. The install needs repair (run the Hermes installer or hermes doctor)."; log "$FINAL_MSG"; exit 3; }
 
 # Run FROM the install root: `hermes update` resolves the tree it mutates
--- a/tests/test_desktop_update_shim_progress.py
+++ b/tests/test_desktop_update_shim_progress.py
@@ -130,11 +130,11 @@ exit "$(cat "$HERMES_TEST_EXITS.$n" 2>/dev/null || echo 0)"
 """
 
 
-def _run_handoff(tmp_path, exits: dict[int, int]) -> list[dict]:
+def _run_handoff(tmp_path, exits: dict[int, int], venv_dir: str = "venv") -> list[dict]:
     """Run the real hand-off end to end; return the stage seen at each call."""
     install_root = tmp_path / "hermes-agent"
-    (install_root / "venv" / "bin").mkdir(parents=True)
-    hermes = install_root / "venv" / "bin" / "hermes"
+    (install_root / venv_dir / "bin").mkdir(parents=True)
+    hermes = install_root / venv_dir / "bin" / "hermes"
     hermes.write_text(FAKE_HERMES)
     hermes.chmod(0o755)
 
@@ -197,3 +197,14 @@ def test_retry_gate_publishes_a_distinct_stage(tmp_path):
         "Updating code and dependencies",
         "Retrying update",
     ]
+
+
+@requires_posix_handoff
+def test_dot_venv_layout_is_not_told_it_needs_repair(tmp_path):
+    """Shared-venv checkouts (doctor-verified healthy) ship only `.venv`, no
+    `venv` directory. The hand-off must still find and run `hermes` instead
+    of aborting with "the install needs repair"."""
+    stages = _run_handoff(tmp_path, {1: 0}, venv_dir=".venv")
+
+    assert len(stages) == 1
+    assert stages[0]["message"] == "Updating code and dependencies"


Enough1122:

**Overall:** Fallback to .venv for posix desktop-update hand-off.

**What it does**
- `scripts/desktop-update/posix.sh` checks `$INSTALL_ROOT/venv/bin/python3` then falls back to `$INSTALL_ROOT/.venv/bin/python3` for browser detection, and similarly for `HERMES_BIN` (`venv/bin/hermes` then `.venv/bin/hermes`) before aborting with "needs repair". Mirrors `managed_uv.py::_default_live_venv()` precedence.
- Test extends `_run_handoff` with `venv_dir` param and adds `test_dot_venv_layout_is_not_told_it_needs_repair` proving .venv-only checkout succeeds.

**Non-blocking notes**
- Fallback uses `[ -x "${py:-/nonexistent}" ]` guard — safe against empty INSTALL_ROOT.

No functional issues found.

monerostar (commented):
Ubuntu 26.04 on linux-5800x.

Ran the real posix.sh harness from this PR tree (cac35ba79):

  tests/test_desktop_update_shim_progress.py::test_dot_venv_layout_is_not_told_it_needs_repair
  tests/test_desktop_update_shim_progress.py::test_retry_gate_publishes_a_distinct_stage

2 passed in 3.12s. The .venv-only fake install got through the repair gate and published the "Updating code and dependencies" stage.

This box's git install still uses ~/.hermes/hermes-agent/venv/bin/hermes (no .venv), so the new fallback would not fire here. The harness covers the shared-venv layout the bug is about.

Looks good.

PR #98395, opened 2026-08-30
refactor(skills): simplify development workflow contracts

topolar:
## Summary
- make TDD risk-proportional while retaining RED/GREEN discipline
- separate root-cause debugging from implementation workflow
- make code review and simplification bounded and proportional
- remove stale workflow references and update generated docs

## Verification
- 1,651 skill tests passed
- 60 skill-manager tests passed
- focused workflow contracts: 5 passed
- git diff --check passed
- independent blocker review: PASS


liyangbing:
<h3>Review: make the contract tests observe the generated boundary</h3><p>The risk-proportional workflow is a good direction, and the revised <code>SKILL.md</code> files now state useful completion criteria. One remaining blind spot is that <code>tests/skills/test_development_workflow_skill_contracts.py</code> mostly proves that phrases occur somewhere in normalized prose. That can stay green if a sentence is moved to the wrong section, a generated page drifts from its source, or an important negative rule is weakened while the marker text remains.</p><h3>Suggested contract additions</h3><p>1. Add a source-to-generated parity check for each changed skill (or invoke the repository's generator and compare the resulting files). This should cover frontmatter fields, headings, procedure order, and the published skill page, not just the catalog entry.<br>2. Assert the decisive negative boundaries at the section level: an adapted TDD path must name its alternative evidence before editing; review must not become a delivery action; and the three-reviewer limit must not be paired with overlapping writers. A whole-document substring check cannot distinguish those cases.<br>3. Add one fixture per workflow that fails for the old contract and passes for the new one: a no-RED documentation/config change, a low-risk review that should stop before commit/push, and two proposed writers with overlapping files. Keep these as policy fixtures rather than snapshots of every sentence.</p><p>I would also make the helper resolve an exact expected path instead of relying only on <code>glob</code>; that turns an accidental duplicate or relocated skill into a clear test failure. The existing 1,651-skill and focused workflow counts are useful, but the PR's strongest promise is synchronization and enforcement across the source and generated surfaces, so those are the boundaries I would protect before marking this ready.</p>

PR #98396, opened 2026-08-30
fix(config): honor provider-level custom_providers context_length on /model switch

### LINKED ISSUE #98387: [Bug] /model switch drops the provider-level custom_providers context_length override — get_custom_provider_context_length() only reads models.<id>.context_length
## Bug Description

On a **custom provider** (self-hosted new-api relay in my case, Hermes v2026.8.3-162-g1be70d6354), the provider-level `context_length` set under `custom_providers[]` is silently ignored on the `/model` switch path, so the context window falls back to the hardcoded catalog (`"glm": 202752`) even though the user configured 1M.

Related but distinct from #97595 / #89500: those are about the *catalog missing glm-5.3*. This report is about the *user's own config override being dropped* on the switch path — which will keep biting for every relay-hosted model no matter how many entries get added to the hardcoded dict.

## Repro

Config (relevant parts):

```yaml
model:
  default: glm-5.3-flash
  provider: custom
  base_url: http://<relay>/v1
  context_length: 1000000
custom_providers:
  - name: my-relay
    base_url: http://<relay>/v1
    model: glm-5.3-flash
    context_length: 1000000     # provider-level override
    models:
      other/model:
        context_length: 1000000 # per-model override
```

Direct repro of the two paths (fresh process, real config loaded):

```python
from hermes_cli.config import load_config, get_custom_provider_context_length, get_compatible_custom_providers
config = load_config()
m = config['model']
cps = get_compatible_custom_providers(config)

# cold-start path (passes model.context_length) -> correct
get_model_context_length(m['default'], base_url=m['base_url'],
    config_context_length=m['context_length'], provider=m['provider'], custom_providers=cps)
# => 1,000,000

# /model switch path (clears _config_context_length, then re-derives):
get_custom_provider_context_length(model='glm-5.3-flash',
    base_url='http://<relay>/v1', custom_providers=cps)
# => None   <-- provider-level context_length: 1000000 is NOT consulted

get_model_context_length('glm-5.3-flash', base_url=m['base_url'],
    provider=m['provider'], custom_providers=cps, config_context_length=None)
# => 202,752  (hardcoded "glm" fallback; relay /v1/models carries no context_length)
```

The relay is new-api (X-New-Api-Version response header visible); its `/v1/models` returns no `context_length` field, the Ollama `/api/show` probe 404s, and custom endpoints skip models.dev — so resolution lands on the hardcoded catalog.

## Root Cause

1. `agent/agent_runtime_helpers.py` (`switch_model`) deliberately clears `agent._config_context_length = None` ("so the new model's actual context window is resolved ... instead of inheriting the stale value") — reasonable intent.
2. It then re-derives the intent via `get_custom_provider_context_length()` (`hermes_cli/config.py`), but that helper **only** reads `custom_providers[i].models.<model>.context_length`. The provider-level `context_length` key documented alongside it is never consulted there.
3. `get_model_context_length()`'s own step-0b (`custom_providers` per-model override, guarded by the same helper) therefore also misses, and the chain falls through to the hardcoded catalog.

So after any `/model` switch to/from the relay model, the compressor + banner show `Context: 203K` and compaction triggers at ~50% of 202K on a model that actually holds 1M. Cold start with the same model as `model.default` is correct (1M) because the startup path reads `model.context_length` directly.

## Fix Suggestion

In `get_custom_provider_context_length()`, after the per-model `models.<id>.context_length` lookup misses, fall back to the matched entry's **provider-level** `context_length` (and optionally `model.context_length` scoped to the same `base_url` at the call sites). This keeps the existing route-mismatch guards (`should_clear_context_pin`) intact while making the override actually reach the switch path.

A structural note echoed from #89500: hardcoded catalog entries will always lag relay-hosted model names — honoring user config overrides consistently across all paths is the durable fix.

Happy to open a PR if maintainers agree with the approach.

liuhao1024:
## What does this PR do?

`get_custom_provider_context_length()` only consulted `custom_providers[i].models.<model>.context_length`. When that per-model key was absent, the entry-level `context_length` declared on the same provider entry (a documented, accepted key — it is in the normalizer's known keys and has a camelCase alias) was silently dropped. The `/model` switch re-derivation in `agent_runtime_helpers.switch_model` deliberately clears `_config_context_length` and re-derives intent through this helper, so every switch to a relay-hosted model fell through to the hardcoded catalog (e.g. `"glm": 202752`) even though the user configured an explicit 1M override — cold start with the same model resolved correctly via `model.context_length`.

This PR makes the helper fall back to the matched entry's provider-level `context_length` when the per-model lookup misses. Per-model values keep priority, and invalid entry-level values (<=0 or non-int) keep falling through to the normal resolution chain, so no new failure mode is introduced. All five call sites listed in the helper's docstring (startup, `/model` switch, `/model` display, `/info` display, `get_model_context_length` step 0b) benefit from the single-function fix.

## Related Issue

Fixes #98387

## Type of Change

## Changes Made

- `hermes_cli/config.py` `get_custom_provider_context_length()`: when the per-model `models.<id>.context_length` lookup misses (no `models` dict, model not listed, or key absent), fall back to the matched entry's provider-level `context_length` instead of skipping the entry; docstring updated to describe the fallback
- `tests/hermes_cli/test_custom_provider_context_length.py`: added `TestEntryLevelContextLengthFallback` (reporter's exact config shape, no-models-dict entry, per-model priority over entry-level, invalid entry-level values still ignored, route isolation) plus a step-0b end-to-end case where the entry-level override wins over the catalog fallback with `config_context_length=None`

## How to Test

1. Offline repro of the issue's exact shape (fresh process, no live endpoint needed):
   ```python
   from hermes_cli.config import get_compatible_custom_providers, get_custom_provider_context_length
   config = {
       "model": {"default": "glm-5.3-flash", "provider": "custom", "base_url": "http://relay.local/v1", "context_length": 1000000},
       "custom_providers": [{"name": "my-relay", "base_url": "http://relay.local/v1", "model": "glm-5.3-flash", "context_length": 1000000, "models": {"other/model": {"context_length": 1000000}}}],
   }
   cps = get_compatible_custom_providers(config)
   print(get_custom_provider_context_length("glm-5.3-flash", "http://relay.local/v1", cps))
   ```
   Observed result: before this PR returns `None` (switch path then lands on 202,752); after this PR returns `1000000`.
2. `python -m pytest tests/hermes_cli/test_custom_provider_context_length.py -v` — Observed result: 15 passed (9 pre-existing + 6 new).
3. `python -m pytest tests/hermes_cli/test_custom_provider_context_length.py tests/hermes_cli/test_model_switch_context_display.py tests/hermes_cli/test_custom_provider_model_switch.py tests/hermes_cli/test_model_switch_custom_providers.py tests/hermes_cli/test_apply_model_switch_result_context.py tests/run_agent/test_switch_model_context.py tests/hermes_cli/test_main_model_custom_provider_normalization.py tests/hermes_cli/test_custom_provider_normalize_no_mutate.py tests/agent/test_model_metadata.py -q` — Observed result: 228 passed.

### Code

- [x] I've run `pytest tests/ -q` and all tests pass (targeted + adjacent surfaces: 243 tests passed across 10 files)
- [x] I've tested on my platform: macOS 15 (arm64), Python 3.11.15

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — docstring in `get_custom_provider_context_length` now describes the fallback; the entry-level key was already an accepted config key
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — or N/A (no new key; existing key now honored)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — pure-Python dict lookup, no platform-specific behavior

## Screenshots / Logs

Offline repro on the fix branch (issue's exact config shape):

```
cold-start path      : 1000000
switch helper        : 1000000
switch path (fixed)  : 1000000
OK — #98387 repro now resolves to 1M on the switch path
```

### DIFF
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -1891,11 +1891,12 @@ def get_custom_provider_context_length(
     custom_providers: Optional[List[Dict[str, Any]]] = None,
     config: Optional[Dict[str, Any]] = None,
 ) -> Optional[int]:
-    """Look up a per-model ``context_length`` override from ``custom_providers``.
+    """Look up a ``context_length`` override from ``custom_providers``.
 
     Matches any entry whose normalized route identity equals ``base_url`` and
     returns ``custom_providers[i].models.<model>.context_length`` if present and
-    valid.  Returns ``None`` when no override applies.
+    valid; when the per-model key is absent, falls back to the matched entry's
+    provider-level ``context_length``.  Returns ``None`` when neither applies.
 
     This is the single source of truth for custom-provider context overrides,
     used by:
@@ -1933,12 +1934,19 @@ def get_custom_provider_context_length(
         if not entry_url or entry_url != target_url:
             continue
         models = entry.get("models")
-        if not isinstance(models, dict):
-            continue
-        model_cfg = models.get(model)
-        if not isinstance(model_cfg, dict):
-            continue
-        raw_ctx = model_cfg.get("context_length")
+        model_cfg = models.get(model) if isinstance(models, dict) else None
+        raw_ctx = (
+            model_cfg.get("context_length")
+            if isinstance(model_cfg, dict)
+            else None
+        )
+        if raw_ctx is None:
+            # No per-model override for this id: honor the provider-level
+            # ``context_length`` declared on the entry itself. Without this
+            # fallback the /model-switch re-derivation — which consults this
+            # helper instead of ``model.context_length`` — silently drops the
+            # user's override and lands on the hardcoded catalog (#98387).
+            raw_ctx = entry.get("context_length")
         if raw_ctx is None:
             continue
         try:
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -1896,7 +1896,11 @@ def get_custom_provider_context_length(
     Matches any entry whose normalized route identity equals ``base_url`` and
     returns ``custom_providers[i].models.<model>.context_length`` if present and
     valid; when the per-model key is absent, falls back to the matched entry's
-    provider-level ``context_length``.  Returns ``None`` when neither applies.
+    provider-level ``context_length``, provided the entry either declares no
+    ``model`` pin or its pin equals ``model`` (a pinned entry is scoped to its
+    own id, so sibling models on the same route cannot inherit it).  Values are
+    accepted only as true positive integers — ``bool`` is rejected because it
+    is an ``int`` subclass in Python.  Returns ``None`` when neither applies.
 
     This is the single source of truth for custom-provider context overrides,
     used by:
@@ -1946,8 +1950,19 @@ def get_custom_provider_context_length(
             # fallback the /model-switch re-derivation — which consults this
             # helper instead of ``model.context_length`` — silently drops the
             # user's override and lands on the hardcoded catalog (#98387).
+            # An entry that pins a specific ``model`` is scoped to that id:
+            # its entry-level value must not leak into a sibling model that
+            # merely shares the same normalized route.
+            declared = entry.get("model")
+            if isinstance(declared, str) and declared and declared != model:
+                continue
             raw_ctx = entry.get("context_length")
         if raw_ctx is None:
+            continue
+        if isinstance(raw_ctx, bool):
+            # bool is an int subclass: int(True) == 1 would silently turn a
+            # true/false typo into a 1-token (or 0) context window instead
+            # of falling through to the catalog chain.
             continue
         try:
             ctx = int(raw_ctx)
--- a/tests/hermes_cli/test_custom_provider_context_length.py
+++ b/tests/hermes_cli/test_custom_provider_context_length.py
@@ -52,6 +52,86 @@ class TestGetCustomProviderContextLength:
         assert get_custom_provider_context_length("m", "http://x", []) is None
 
 
+class TestEntryLevelContextLengthFallback:
+    """The provider-level ``context_length`` key must be honored when the
+    per-model ``models.<id>.context_length`` lookup misses (#98387).
+
+    The ``/model`` switch re-derivation consults this helper instead of the
+    top-level ``model.context_length``, so dropping the entry-level key
+    made every switch to a relay-hosted model fall through to the hardcoded
+    catalog even though the user configured an explicit override.
+    """
+
+    def test_entry_level_used_when_model_not_in_models_dict(self):
+        """The reporter's shape: entry-level override set, but ``models:``
+        only lists *other* models — the queried id must still resolve."""
+        custom = [
+            {
+                "base_url": "https://relay.example.invalid/v1",
+                "model": "glm-5.3-flash",
+                "context_length": 1_000_000,
+                "models": {"other/model": {"context_length": 1_000_000}},
+            }
+        ]
+        assert (
+            get_custom_provider_context_length(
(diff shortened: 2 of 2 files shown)


liyangbing:
<h3>Review: add same-route model isolation and strict integer validation</h3><p>The new fallback is correctly tested for a missing <code>models</code> map, per-model-over-entry precedence, route isolation, and top-level config precedence. Two adjacent cases are worth pinning before merge because the fallback now applies when the per-model lookup misses:</p><p>1. Put two custom-provider entries on the same normalized <code>base_url</code>, each with a different <code>model</code> and a different entry-level <code>context_length</code>, then query both models in both list orders. The entry-level value should follow the matching model entry, or the documented precedence should explicitly say that a shared route makes the value global. Otherwise the first matching entry can leak its context limit into a sibling model during <code>/model</code> switch.<br>2. Add <code>True</code> and <code>False</code> to the invalid entry-level values. In Python, <code>bool</code> is an <code>int</code>; accepting it would turn a configuration typo into a context window of 1 or 0 instead of falling through. The validation contract should use an explicit non-boolean integer check, with a positive bound.</p><h3>End-to-end assertion</h3><p>For the same-route fixture, exercise <code>get_model_context_length</code> with <code>config_context_length=None</code> and a catalog fallback for each model, then assert the selected entry-level value survives both cold startup and the switch path. That would prove the fix is not only correct for the helper's local return value but also preserves model identity at the caller that motivated #98387.</p><p>These are small additions, but they protect the new fallback's two implicit assumptions: which provider entry owns an override, and what counts as a valid numeric limit.</p>

liuhao1024:
Thanks @liyangbing — both cases are now pinned, and the first one turned out to be a real leak, not just a missing test.

**1. Same-route model isolation** (a44b5cc): I implemented the isolation rather than documenting the route-global behavior, because the leak you described was real — the first matching entry would hand its entry-level limit to a sibling model on the same normalized `base_url`. The fallback now only applies when the entry declares no `model` pin, or its pin equals the queried id:

```python
declared = entry.get("model")
if isinstance(declared, str) and declared and declared != model:
    continue
```

Two pinned entries sharing a route each keep their own limit, verified in both list orders. The counterpart semantics is now explicit in the docstring: an entry *without* a pin claims the whole route (route-global), and a pinned sibling no longer shadows a later unpinned entry for other models.

**2. Strict integer validation**: `True`/`False` are in the invalid-value matrix now, and the helper rejects `bool` explicitly before the `int()` coercion (bool is an int subclass, so `int(True) == 1` would have turned a typo into a 1-token window instead of falling through). The check sits on the shared validation chain, so it covers per-model values too.

**3. End-to-end assertion**: added `test_same_route_entry_level_keeps_model_identity_through_resolver`, which drives `get_model_context_length` with `config_context_length=None` and every catalog probe mocked, once per model of the same-route fixture — asserting each id resolves to its own entry-level value through the full chain (proving the caller keeps model identity, not just the helper return) — plus the cold-startup shape where an explicit top-level `model.context_length` still wins.

All 20 tests in `tests/hermes_cli/test_custom_provider_context_length.py` and the 6 in `tests/run_agent/test_switch_model_context.py` pass.

teknium1:
#98387 106976

Issue #98378, opened 2026-08-30
[Bug]: An unrelated service in STOP_PENDING aborts the whole update


INTEGRITY2077:
## Summary

On Windows, an unrelated service sitting in `STOP_PENDING` can abort the whole
update. If that service's PID happens to appear among the gateway process's
ancestors, SCM discovery raises and the update stops — even though the service
has nothing to do with Hermes and will finish stopping on its own moments
later.

I hit this with `BcastDVRUserService`, the Windows Game DVR per-user service,
which cycles through `STOP_PENDING` routinely.

## Environment

- OS: Windows 11 Pro 26200
- Hermes desktop app, updating from the in-app update button
- `hermes-agent` at `%LOCALAPPDATA%\hermes\hermes-agent`

## Where it happens

`hermes_cli/gateway.py` treats any non-`running`, non-`stopped` service as
indeterminate:

```python
# hermes_cli/gateway.py:1087-1094
if service_status == "stopped":
    continue
if service_status != "running":
    if service_pid > 0:
        indeterminate_services_by_pid.setdefault(service_pid, []).append(
            (service_name, service_status)
        )
    continue
```

and then raises if such a PID is an ancestor of the gateway process:

```python
# hermes_cli/gateway.py:1112-1119
ancestor_pids = [int(parent.pid) for parent in gateway_process.parents()]
for pid in ancestor_pids:
    indeterminate_services = indeterminate_services_by_pid.get(pid, [])
    if indeterminate_services:
        service_name, service_status = indeterminate_services[0]
        raise RuntimeError(
            f"SCM service {service_name} has indeterminate status: "
            ...
        )
```

Note that the code elsewhere already knows `STOP_PENDING` is a transient state
worth waiting out — `hermes_cli/update_cmd.py:5802` is documented as
"Restore a service after an uncertain stop, including STOP_PENDING", and
`tests/hermes_cli/test_update_concurrent_quarantine.py:416` covers
`test_restore_windows_gateway_service_waits_out_stop_pending`. The discovery
path above does not wait; it raises immediately.

## Steps to reproduce

The exact ancestry is timing-dependent, so this is the shape of it rather than
a deterministic recipe:

1. On Windows, have a per-user service that cycles through `STOP_PENDING`.
   `BcastDVRUserService_*` does this on its own; it can also be pushed there
   with `sc stop BcastDVRUserService_<suffix>`.
2. While that service is in `STOP_PENDING`, run the update.
[...]

PR #97929, opened 2026-08-29
fix(update): unrelated transitional Windows services no longer abort updates (salvage of #96381)

kshitijk4poor:
## Summary

Salvage of PR #96381 by @fangliquanflq onto current `main`, authorship preserved.

A single unrelated Windows service stuck in a transitional SCM state (paused, START_PENDING, STOP_PENDING — e.g. FontCache3.0.0.0, cortsmartserver) no longer aborts every Desktop/CLI update with "SCM service enumeration failed" (#96860, #97856, #96360). `find_windows_gateway_services()` now records transitional services instead of raising during enumeration, and stays **fail-closed exactly where it matters**: if a transitional service PID sits in a validated gateway's ancestor chain, the update still aborts with the original error.

Cluster resolution (4 competing PRs):
- **#96381 (this) wins**: precise semantics — skip unrelated transitional services, fail closed on gateway-ancestry ones; smallest correct diff (+49/-6); both regression directions tested (unrelated STOP_PENDING passes; ancestor STOP_PENDING raises).
- #97238: skips ALL non-running services including a gateway's own transitional host — loses the fail-closed guarantee 790e1eb6bd deliberately added.
- #96398: same semantics as #96381 plus PID create-time identity re-checks for transitional services (+265/-5) — heavier variant of the same idea; the identity re-check duplicates the check already done on the gateway side of the join.
- #97220: different bug (Task Scheduler ownership misattribution) — NOT closed as a duplicate; triaged separately.

## Changes
- `hermes_cli/gateway.py`: transitional services collected into `indeterminate_services_by_pid` instead of raising; raise deferred to ancestry join
- `tests/hermes_cli/test_gateway.py`: unrelated `stop_pending` service tolerated; transitional ancestor still fail-closed

## Validation
| | result |
|---|---|
| tests/hermes_cli/test_gateway.py | 35 passed, 3 skipped |
| mutation (gateway.py reverted to main) | 2 failed → restored: 35 passed |

Closes #96381
Fixes #96860
Fixes #97856
Fixes #96360
Refs #97238, #96398

PR #98397, opened 2026-08-30
fix(update): wait out transient STOP_PENDING/START_PENDING in SCM gateway discovery

lEWFkRAD:
## What

On Windows, an unrelated service sitting in `STOP_PENDING`/`START_PENDING` could abort the entire update via SCM gateway discovery. `find_windows_gateway_services()` now waits out the transient state on that one service instead of raising immediately, closing #98378.

## Root cause

`hermes_cli/gateway.py::find_windows_gateway_services()` classified every service status that is neither `running` nor `stopped` — including the *transient* `stop_pending`/`start_pending` — as indeterminate, and hard-raised if such a service's PID appeared anywhere in a validated gateway's ancestor chain. Per-user services like `BcastDVRUserService_*` cycle through `STOP_PENDING` routinely, and svchost ancestry sharing makes an unrelated pending service's PID show up in that chain by coincidence. The update then dies with `SCM service <name> has indeterminate status: stop_pending` even though the service settles on its own within moments.

The codebase already treats these states as wait-and-retry in the restore path (`update_cmd._restore_windows_gateway_service`); the discovery path just never got the same treatment.

## Fix

When the enumeration sees `stop_pending`/`start_pending` for a service with a live PID, re-poll **that one service** (0.2 s interval, ~10 s bound — nothing else in the scan grows a wait):

- settles **stopped** → skipped (a stopped service can't supervise a live gateway)
- settles **running** → registered normally, with the re-read status/pid flowing into the ownership checks
- service **deleted** mid-wait → skipped, same as the existing `FileNotFoundError` handling
- still pending / unknown at the deadline, or any genuinely-uninspectable state → raises exactly as before

Everything fail-closed stays fail-closed: empty names, running-without-PID, shared-host ambiguity, identity-change, and non-transitional indeterminate statuses are untouched (the pre-existing `test_find_windows_gateway_services_rejects_transitional_ancestor` now pins the still-pending-past-deadline case).

Contributes to #98378 (direction 1 — wait it out — from the issue's suggested directions). Not claiming `Closes`: the issue's repro is timing-dependent, so final confirmation should come from the reporter's real update attempts.

## Tests

`tests/hermes_cli/test_gateway.py`:
- `..._waits_out_stop_pending_to_stopped` — transient ancestor settles `stopped` → discovery completes, no raise
- `..._stop_pending_settling_running_is_registered` — settles `running` → registered with full service/PID/identity contract
- `..._rejects_transitional_ancestor` — updated to pin a *fake clock* (no real sleeping), so it now genuinely exercises the bounded wait and still fail-closes

```
uv run --with pytest --with python-dotenv python -m pytest \
  tests/hermes_cli/test_gateway.py tests/hermes_cli/test_update_concurrent_quarantine.py \
  tests/hermes_cli/test_windows_gateway_cold_start_desktop_lifecycle.py -q
68 passed, 5 skipped, 3 failed
```

The 3 failures (`TestReapUnsupervisedGatewayOrphans*`, `TestReaperCandidateIsSupervisorOwned`) fail identically on a pristine `origin/main` worktree at 26350357d7 on this machine — pre-existing, environment-dependent, and untouched by this diff.

### DIFF
--- a/hermes_cli/gateway.py
+++ b/hermes_cli/gateway.py
@@ -1061,7 +1061,35 @@ def find_windows_gateway_services(
         if profile_processes is None:
             profile_processes = find_profile_gateway_processes(strict=True)
         service_names_by_pid: dict[int, set[str]] = {}
-        indeterminate_services_by_pid: dict[int, list[tuple[str, object]]] = {}
+        # Transient states (STOP_PENDING/START_PENDING) are re-polled instead of
+        # raising immediately — an unrelated service cycling through them must
+        # not abort the update (#98378), mirroring update_cmd's
+        # _restore_windows_gateway_service. Genuine errors still fail closed.
+        transient_services_by_pid: dict[int, list[tuple[str, object]]] = {}
+
+        def _re_poll_service_status(current_service, current_status):
+            """Return a stable status, waiting out the transient state.
+
+            Polls only this one service (never the whole SCM) at 0.2s
+            intervals until a ~10s deadline. ``None`` means the service is
+            gone or became re-inspectable as stable/unknown at the deadline;
+            the caller then re-reads it through the normal path.
+            """
+            if current_status not in ("stop_pending", "start_pending"):
+                return current_status
+            deadline = time.monotonic() + 10.0
+            while time.monotonic() < deadline:
+                time.sleep(0.2)
+                try:
+                    polled_status = current_service.status()
+                except FileNotFoundError:
+                    return None
+                except Exception:
+                    return current_status
+                if polled_status not in ("stop_pending", "start_pending"):
+                    return polled_status
+            return None
+
         for service in psutil_module.win_service_iter():
             try:
                 if all(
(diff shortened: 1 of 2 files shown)


liyangbing:
<h3>Review: bound the discovery pass, not only each service wait</h3><p>The per-service retry is a sensible containment for the reported false positive, and the fail-closed deadline behavior is preserved. One timing contract is still ambiguous: if enumeration returns several unrelated services in <code>STOP_PENDING</code>/<code>START_PENDING</code>, a roughly 10-second wait for each can make the total gateway discovery delay grow as <code>N × 10s</code>. The PR says that nothing else in the scan grows a wait, so it would help to make that an explicit invariant rather than an implementation assumption.</p><h3>Suggested regression matrix</h3><p>1. Two or more pending services, including one unrelated service before the actual gateway service, with a fake monotonic clock: assert the discovery pass has a documented global upper bound and still reaches the gateway candidate.<br>2. Exercise both <code>STOP_PENDING</code> and <code>START_PENDING</code>, plus a service that is deleted while pending.<br>3. Let a pending service settle to <code>running</code> with a changed PID, and separately let its name/status/PID change between the final re-read and the ancestor ownership check. Registration should use one coherent, freshly validated identity snapshot or fail closed; it must not carry the original pending PID into the ownership decision.<br>4. Keep a pending service past the deadline and assert the error includes the service identity and observed state without masking a later unrelated candidate.</p><p>The existing fake-clock test is a good foundation. Adding a multi-service case and an identity-change case would protect the two races this fix introduces: cumulative startup latency and a status/PID snapshot becoming stale during the wait.</p>

alt-glitch:

Duplicate of merged #97929. Current main already repairs the unrelated transitional-SCM-service update failure with an ownership-first, fail-closed approach.

lEWFkRAD:
CI red on `tests/gateway/test_goal_continuation_drain.py::test_runner_goal_hook_enqueues_into_the_key_the_adapter_drains` (Slack continuation-key assertion) is unrelated to this diff — this PR touches only `hermes_cli/gateway.py` SCM service discovery and `tests/hermes_cli/test_gateway.py`; the failing test exercises message-queue key routing in `gateway/runner`, untouched here, and none of the changed functions appear anywhere in its call path. Affected-tests CI also labels it 'unrelated' in the scan rows.

PR #98401, opened 2026-08-30
test(cli): cover azure_detect pure credential and path helpers

salch-cred:
## Why

`azure_detect` routes Azure Foundry endpoints to the right API mode (OpenAI vs Anthropic). Its pure helper functions — credential resolution, auth-header application, path sniffing, model-id extraction — had zero direct test coverage.

## What

`tests/hermes_cli/test_azure_detect_helpers.py`:

- `_resolve_credential`: API key string, empty key, None, callable token provider, callable api_key, provider wins over string, failing callable degrades to `(None, entra_id)`.
- `_apply_auth_headers`: api_key sends both `api-key` and `Authorization` headers; entra_id sends only Bearer; no token sets no headers.
- `_looks_like_anthropic_path`: path ends with `/anthropic`, contains `/anthropic/` segment, OpenAI-style URL false, case-insensitive, bare host false, anthropic-only-in-hostname false.
- `_strip_trailing_v1`: strips `/v1` with/without slash, no-op when absent, mid-path `/v1/` not stripped.
- `_extract_model_ids`: standard OpenAI shape, empty data, missing data key, non-dict payload, item missing id skipped, fallback to `model`/`name` keys.
- `DetectionResult` defaults.

### DIFF
--- a/tests/hermes_cli/test_azure_detect_helpers.py
+++ b/tests/hermes_cli/test_azure_detect_helpers.py
@@ -1,0 +1,187 @@
+"""Behavioral tests for the pure helper functions in hermes_cli.azure_detect.
+
+The public `detect()` function makes live HTTP calls and is not tested here.
+These tests cover the pure parsing/routing helpers that have no I/O.
+"""
+
+import pytest
+
+from hermes_cli.azure_detect import (
+    DetectionResult,
+    _apply_auth_headers,
+    _extract_model_ids,
+    _looks_like_anthropic_path,
+    _resolve_credential,
+    _strip_trailing_v1,
+)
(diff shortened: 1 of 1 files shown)


Enough1122:

**Overall:** Adds tests for azure_detect credential/path helpers.

**What it does**
- New `tests/hermes_cli/test_azure_detect_helpers.py` covers `_resolve_credential` (api key string/empty/None, callable token provider, callable api_key, precedence, failing callable), `_apply_auth_headers` (api_key sends both headers, entra_id only Bearer), plus path sniffing and model-id extraction — pure helpers, live HTTP `detect()` intentionally not tested.

No issues found.

Issue #98394, opened 2026-08-30
Desktop renderer permanent re-render loop — entire chat content briefly disappears and reappears


aSchaber2026:
# Bug: Desktop renderer permanent re-render loop — entire chat content briefly disappears and reappears

## Summary

The Hermes Desktop app's Electron renderer process sustains **30–65% CPU** in
idle (no active turn), causing the entire chat thread to briefly unmount and
remount on a visible cycle. The user sees the whole text area flash empty
(~1 frame) then return, accompanied by the scroll-to-bottom arrow popping in.
The backend is quiet — this is a frontend-only render loop.

## Environment

| | |
|---|---|
| Hermes Agent | v0.20.6 (2026.8.27), commit `52e5e7c03` |
| Desktop app | v0.17.0 (CFBundleVersion), Electron 40.10.2 |
| OS | macOS 26.6.2 (Build 25G83) |
| Hardware | Apple M5, 32 GB RAM (Mac17,3) |
| Profiles | 4 (`default`, `haus`, `haus-technik`, `recherche`) |
| Session store | 326 sessions, 93,574 messages, 841 MB state.db |
| Active session | 183 messages, 377k input + 42k output tokens |

## Reproduction

1. Open Hermes Desktop with a multi-profile setup (4 profiles) and a large
   session store (>300 sessions, >90k messages).
2. Open a session with 180+ messages and a long tool-turn history.
3. Observe the renderer process CPU in Activity Monitor: stays 30–65% even
   while no turn is running and no user input is happening.
4. Watch the chat area: the entire thread content briefly disappears and
   reappears on a ~1–5 second cycle. The scroll-to-bottom arrow
   (`ScrollToBottomButton`, `codicon arrow-down`) pops in each time.

### Renderer CPU (idle, no active turn)

5 samples, 2 seconds apart, PID 83011 (Hermes Helper Renderer):

```
31.8%  48.8%  42.4%  55.9%  41.7%   (RSS ~700–736 MB)
```

GPU process (PID 82997) simultaneously:

```
35.2%  26.3%  31.9%  32.2%  32.7%
```

Backend Python process (PID 83277, `hermes_cli.main serve`) for comparison:

```
0.5%  9.7%  7.6%  4.7%  0.8%   (only spikes during active turns)
```

### Backend logs — quiet

- `sessions.changed` WS broadcasts: **0** in the 20+ minutes after the last
  desktop restart (07:45–08:10). The change watcher
[...]

PR #98408, opened 2026-08-30
fix(desktop): stop the HUD transcript-band probe once the viewport mounts

chelsealong:
Fixes one concrete, verifiable defect named in #98394 ("Desktop renderer permanent re-render loop — entire chat content briefly disappears and reappears").

## What's actually wrong here

The issue lists five `setInterval`-based timers as suspects for the sustained idle renderer CPU / re-render behavior. I checked each one against current `main`:

| Timer | Status on current main |
|---|---|
| `app/agents/index.tsx:200` `setNowMs` (500ms) | Already gated on `active > 0 && visible` (`usePaneVisible()`) — does not fire when idle or the panel isn't visible. Not a bug. |
| `contrib/runtime-loader.ts:590` disk plugin scan (5s) | Already gated on `document.visibilityState === 'visible'`. Not a bug. |
| `plugins/kanban/ui.tsx:101` `useTicking` force-render (5s) | Scoped to a single mounted card's own subtree, not a global re-render. Not a bug. |
| `app/hud/hud-shell.tsx:269` edge-flip poll (300ms) | Intentional continuous poll — no DOM move event exists, comment says so explicitly. Not a bug. |
| `app/hud/hud-shell.tsx:361` transcript-band "probe" (500ms) | **Actually broken** — see below. |

The transcript-band probe's own comment says: *"The viewport mounts async (lazy chat surface); poll briefly until it exists, then let the ResizeObserver own it."* The code never implemented the "poll briefly" part — the interval called `measure()` unconditionally, forever, for the entire lifetime of the HUD window, regardless of whether the ResizeObserver had already taken over. That's a real, sustained per-window timer running DOM queries + `getBoundingClientRect` + a style write every 500ms indefinitely, exactly the class of defect the issue reports (a HUD-window analog of the main-window timers it audits).

## Fix

Extracted the effect into `useHudTranscriptBand()` (matching this file's existing per-concern hook split — `useHudGlass`, `useHudClickThrough`, `useHudThreadFocus` already live this way) and made the interval check whether the viewport has been found before calling `measure()` again, clearing itself the first tick after it has — so the ResizeObserver takes over exactly as the original comment intended.

## Test

Added `apps/desktop/src/app/hud/transcript-band.test.tsx`, which mounts the hook with an inert `ResizeObserver` stub and fake timers, and asserts the probe keeps firing while the viewport hasn't mounted, fires exactly once more the tick it appears, and then stops (measured via a `getBoundingClientRect` call-count spy, since that's called unconditionally on every `measure()`).

Confirmed the test fails without the fix — reverted the interval to the old unconditional `setInterval(measure, 500)` and reran:

```
❯ |ui| src/app/hud/transcript-band.test.tsx (1 test | 1 failed) 56ms
   × stops polling once the viewport mounts
AssertionError: expected 90 to be 10 // Object.is equality
```

Restored the fix and reran — passes:

```
 Test Files  1 passed (1)
      Tests  1 passed (1)
```

Full `hud/` suite (6 files, includes the new test) still green:

```
 Test Files  6 passed (6)
      Tests  25 passed (25)
```

`npx eslint src/app/hud/hud-shell.tsx src/app/hud/transcript-band.ts src/app/hud/transcript-band.test.tsx` — clean.

`npx tsc -p . --noEmit` (whole project) — clean.

I did not run the entire `apps/desktop` vitest suite (684 test files) end-to-end — it exceeded the time budget of this environment. The change is scoped to one file split into two, with no behavioral change to any other module (`hud-shell.tsx`'s only change is delegating to the extracted hook), so I limited verification to the affected area plus a full-project typecheck.

## Disclosure

This PR was prepared with AI assistance (an autonomous Claude Code session), with the diff, test, and repro reviewed before submission.

### DIFF
--- a/apps/desktop/src/app/hud/hud-shell.tsx
+++ b/apps/desktop/src/app/hud/hud-shell.tsx
@@ -13,9 +13,9 @@ import { useHudClickThrough } from './click-through'
 import { useHudGameOverlay } from './game-overlay'
 import { useHudGlass } from './glass'
 import { useHudGoto, useReportHudSession } from './handoff'
-import { hudTranscriptHeight } from './layout'
 import { hudResizeDirections, useHudResizeHandle } from './resize-handle'
 import { useHudThreadFocus } from './thread-focus'
+import { useHudTranscriptBand } from './transcript-band'
 
 /** How long the transcript lingers at its glanceable opacity — after a turn
  *  lands, or after you let go of the composer — before it goes. This is the ONLY
@@ -38,11 +38,6 @@ const HUD_DIM_MS = Math.round(HUD_FADE_MS * 1.5)
  *  while the last of the text is still going — it reads as the transcript being
  *  drawn down into the bar rather than the two dissolving in lockstep. */
 const HUD_COLLAPSE_MS = Math.round(HUD_FADE_MS * 0.66)
-
-/** Breathing room the sheet keeps above the first row, so the fade has
- *  somewhere to land. Folded into the measured height rather than added in CSS,
- *  so an empty transcript measures a true zero instead of a 12px strip. */
-const HUD_SHEET_OVERHANG_PX = 12
 
 /** Composer on top, transcript always hanging below it — Spotlight's shape,
  *  rather than flipping to follow the screen edge the HUD is parked against. */
@@ -275,6 +270,8 @@ export function HudShell() {
     }
   }, [])
 
+  const rootRef = useRef<HTMLDivElement | null>(null)
+
   // Whether bar + band actually cover the window. Gates the frost, which is
   // native vibrancy and therefore the WINDOW's content view — it fills the whole
   // rectangle and nothing in the page can clip it to the sheet. Whenever the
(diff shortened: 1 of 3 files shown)


Enough1122:

**Overall:** Fixes sustained HUD transcript-band probe; stops permanent re-render loop candidate.

**What it does**
- `apps/desktop/src/app/hud/hud-shell.tsx` extracts continuous ResizeObserver measure/probe logic into new hook `useHudTranscriptBand` (new file `transcript-band.ts`) and removes inline `useEffect` that observed viewport/firstChild forever. New hook presumably gates observation after viewport mounts.
- Removes unused `hudTranscriptHeight` import and constants, adds `rootRef` placement.

**Non-blocking notes**
- Diff truncated; verify new hook cleans up ResizeObserver on unmount and does not re-create observer each render — otherwise leak persists in different form. Correctness likely fine per description of single concrete defect fix.

No blocking issues found from visible diff.

OutThisLife (approved):
LGTM — the transcript-band probe now stops once the viewport exists, matching its own comment.

teknium1:
Salvaged onto current `main` as #100581 with your commit cherry-picked and authorship preserved (clean pick, no conflicts). Your timer audit table also re-verified accurate on today's main — it's quoted in the salvage PR body and in the #98394/#98683 status updates. Thanks for the disciplined sabotage-tested fix; this PR will be closed in favor of the salvage once it merges.

teknium1:
Thanks @chelsealong — landed on main via #100581 with your authorship preserved (clean cherry-pick): the HUD transcript-band probe now stops once the viewport mounts. Closing as superseded by the merged salvage.

Issue #72316, opened 2026-07-27
[Bug]: `_should_treat_stop_as_truncated` false-positives on Ollama Cloud GLM + SSE session stream doesn't propagate `partial` status


Joly0:
### Bug Description

Two compounding bugs cause the Hermes WebUI (nesquena/hermes-webui) to discard or misrender agent responses when using GLM-5.2 on Ollama Cloud:

1. **False-positive truncation detection** — `_is_ollama_glm_backend()` in [`run_agent.py`](https://github.com/NousResearch/hermes-agent/blob/main/run_agent.py#L1595) matches `ollama.com` (the cloud hosted service), not just local Ollama (`:11434`). Combined with the overly strict `_has_natural_response_ending()` heuristic ([`run_agent.py#L1576`](https://github.com/NousResearch/hermes-agent/blob/main/run_agent.py#L1576)), any response not ending with a narrow punctuation whitelist (`. ! ? : ) " ' ] }`, CJK punctuation, or emoji) gets `finish_reason: "stop"` converted to `"length"`, injecting a spurious `[System: Your previous response was truncated by the output length limit...]` continuation prompt into the conversation.

2. **SSE session stream hardcodes `partial: False`** — `_handle_session_chat_stream()` in [`api_server.py`](https://github.com/NousResearch/hermes-agent/blob/main/gateway/platforms/api_server.py#L3552) always sends `"partial": False` in the `assistant.completed` event, ignoring `result.get("partial")`. The WebUI can't detect truncation and renders partial responses incorrectly (showing only the last few lines — the continuation — instead of the full response).

### Steps to Reproduce

1. Configure: `model: glm-5.2`, `provider: ollama-cloud`, `base_url: https://ollama.com/v1`
2. Send a message that triggers tool calls, then produces a text response not ending with `.!?:"\]'}`, CJK punctuation, or emoji (e.g. ends with a newline, a word, or a closing tag that isn't `` ``` ``)
3. Observe: `[System: Your previous response was truncated by the output length limit...]` appears in the session transcript as a user message
4. Agent logs show `finish_reason=stop` (not `length`) — confirming false positive
5. WebUI shows only the last few lines (the continuation) instead of the full response

### Expected Behavior

[...]

PR #98410, opened 2026-08-30
fix(agent): cap Ollama/GLM suspicious-stop continuation loop at one salvage attempt

### LINKED ISSUE #98406: False truncation loop with feedback: _should_treat_stop_as_truncated rewrites correct stop on Ollama Cloud GLM, then the continuation nudge itself causes the next truncation
## Summary

On `glm-5.3-flash:cloud` via Ollama (`http://127.0.0.1:11434/v1`), benign conversational turns are getting killed with:

```
Provider error — Response remained truncated after 4 continuation attempts
```

The root cause is not an actual truncation. `_should_treat_stop_as_truncated()` rewrites a correct `finish_reason="stop"` into `"length"`, and then the 4 continuation retries make things *worse* than the original response would have been. What I found in the logs makes this worse than the existing reports (#72316, #98255, #58087): on GLM 5.3 Flash with reasoning enabled, the model spends its next output budget *reasoning about the synthetic continuation nudge*, so each retry makes another false-positive truncation more likely. The recovery mechanism becomes the failure mode.

## Environment

- hermes-agent at `26350357d7` (desktop app 0.17.0)
- `model.default: glm-5.3-flash:cloud`, `model.base_url: http://127.0.0.1:11434/v1` (Ollama cloud model, local proxy endpoint)
- `reasoning_effort: high` in the session where the screenshot was taken

### 1. The rewrite happens on a *stop*, not a real length cap

From `~/.hermes/logs/desktop.log` (2026-08-30, timestamps are UTC):

```
06:24:48 ⚠️  Treating suspicious Ollama/GLM stop response as truncated
06:24:48 ⚠️  Response truncated (finish_reason='length') - model hit max output tokens
06:25:03 ⚠️  Treating suspicious Ollama/GLM stop response as truncated
06:25:12 ⚠️  Treating suspicious Ollama/GLM stop response as truncated
06:25:19 ⚠️  Treating suspicious Ollama/GLM stop response as truncated
06:25:19 ⚠️  Response still truncated after 4 continuation attempts
```

Ollama reported `stop` four times in a row. Four independent responses, all judged "truncated" by the heuristic, ~9-15s apart — that is not an output-token cap biting four times in a row on a small chat turn, that is the heuristic rejecting normal text endings.

### 2. The continuation nudge itself poisons the next turn

The user-visible consequence in the desktop app was GLM's thinking block leaking into the UI, reasoning about the injected nudge:

> "The user's message is just a system continuation prompt. My previous message was already the final summary. I should not repeat or do anything new — just a brief closing..."

That is `_LENGTH_CONTINUATION_OUTPUT_LIMIT` (`agent/conversation_loop.py:1180`) being consumed as conversation content by a reasoning model. On glm-5.3-flash with `reasoning_effort: high`, the model burns a large slice of its output budget deliberating over a synthetic `[System: ...]` user message, which produces exactly the unpunctuated / mid-thought endings that `_has_natural_response_ending()` rejects. The loop is self-reinforcing:

```
stop (correct) ──rewrite──> length
  └─> inject nudge ──> model reasons about nudge ──> unpunctuated tail
        └─> 4x "still truncated" ──> partial response + error banner
```

### 3. Code path

- `run_agent.py:1807` `_should_treat_stop_as_truncated()` — introduced in `8011aa31ba` ("fix(agent): continue ollama glm truncation replies"). Guard chain: `finish_reason=="stop"` + `_is_ollama_glm_backend()` + a `role:"tool"` message exists in history + visible text ≥ 20 chars + `_has_natural_response_ending()` says no.
- `run_agent.py:1766` `_has_natural_response_ending()` — accepts only `.!?:)]}"'` ``` or emoji endings. Anything else (a bare word, an unclosed fence, a list item) counts as truncated.
- `run_agent.py:1785` `_is_ollama_glm_backend()` — matches `"glm" in model` + `":11434" in base_url`. Ollama cloud models served through the local proxy on 11434 match this even though the actual generation happens on Ollama's hosted endpoint, which reports `finish_reason` correctly.
- `agent/conversation_loop.py:3960-4042` — the 4-attempt continuation loop that appends the nudge and ultimately returns `error: "Response remained truncated after 4 continuation attempts"`.

### 4. Trigger frequency

`grep -c "Treating suspicious" ~/.hermes/logs/desktop.log` → **391 hits** on this install, clustered on glm sessions. The heuristic fires constantly, and every fire risks escalation to the 4-attempt error.

## Why the original fix doesn't fit this case

`8011aa31ba` targets a real historical bug: local Ollama misreporting tool-call stream truncations as `stop`. But this setup is **Ollama cloud** (`glm-5.3-flash:cloud`) — generation runs server-side, and the local 11434 endpoint is just a proxy that faithfully forwards the upstream `finish_reason`. The docstring of `_is_ollama_glm_backend()` itself says the detection exists because of misreporting on local inference; assuming a cloud-backed response is misreported needs more than the port number the proxy happens to listen on.

## Related issues

- #72316 — same false-positive family on Ollama Cloud GLM (flagged `needs-decision`)
- #98255 — `_has_natural_response_ending` rejects every unpunctuated reply
- #58087 — spurious continuation retries on short complete responses

I'm opening this separately because the nudge-induced reasoning exhaustion loop (evidence §2) is a distinct failure mode none of the above document: the recovery prompt itself consumes the output budget and manufactures the next false truncation.

## Suggested direction

Happy to work on a fix. Sketch of what seems sane, ordered by blast radius:

1. **Trust upstream `stop` for cloud models**: if the Ollama model id carries `:cloud` (or the response reports server-side usage stats), skip the stop→length rewrite. The misreport the heuristic guards against is a local-inference artifact.
2. **Require positive evidence of truncation**, not just a missing punctuation mark: e.g. text ending mid-fence, or `completion_tokens == max_tokens` agreement, before rewriting `stop` to `length`.
3. **Make the continuation nudge invisible to reasoning models** (it currently leaks as a `[System: ...]` user turn) or exempt nudge-driven retries from re-triggering the heuristic, otherwise the loop never converges.

Would a PR along line 1 (+ tests) be welcome, or would maintainers rather see 2+3 combined?

0xSteveTheHead:
## Problem

On Ollama-hosted GLM models, `_should_treat_stop_as_truncated` (from #8011aa31ba) rewrites a provider-correct `finish_reason="stop"` into `"length"` whenever the visible text lacks a "natural" punctuation ending. The four-attempt continuation loop then makes things strictly worse for reasoning models: the injected `[System: ...]` continuation nudge itself becomes the topic of the next reasoning pass — the model burns its output budget deliberating over the synthetic message and produces yet another unpunctuated tail, which the heuristic flags again. Net effect on glm-5.3-flash:cloud (`reasoning_effort: high`): four full-budget generations, then the turn dies with

```
Response remained truncated after 4 continuation attempts
```

on what was a complete response. Logs from one desktop install show **391** heuristic firings; #98406 documents the full chain with live evidence, including the model's leaked thinking block reasoning about the nudge itself.

## Fix

Track the heuristic-driven rewrite with a per-API-call flag (`agent._suspicious_stop_rewrite`, consumed and cleared by the truncation handler) and let the continuation loop converge **after one salvage continuation** instead of four:

- first suspicious event → one continuation request (unchanged) — this preserves exactly the tool-tail salvage behavior `8011aa31ba` was written for, and its test still passes
- second suspicious event → accept the stitched partial as the complete turn (`completed=True`, no error banner) instead of two more full-budget retries + hard error
- a genuine `length` cap (flag clean) runs the full 4-attempt loop exactly as before

The convergent path also reframes the exit: no "still truncated after 4 attempts" warning, since the heuristic never saw real provider evidence of a cap.

## Testing

- New `test_ollama_glm_suspicious_stop_converges_after_one_continuation` reproduces the live #98406 shape (tool turn → suspicious stop → suspicious stop after continuation) and asserts convergence: `completed=True`, `error=None`, stitched response delivered
- Existing `test_ollama_glm_stop_after_tools_without_terminal_boundary_requests_continuation` still passes — the salvage-first behavior is preserved
- `tests/run_agent/test_continuation_ceiling_wedge.py` + `test_partial_stream_finish_reason.py`: 27 passed
- `tests/agent/ -k "truncat or continuation or ollama"`: 59 passed, 1 skipped
- Full `tests/run_agent/test_run_agent.py`: 281 passed; the one failure (`TestAnthropicInterruptHandler`) is pre-existing on a clean checkout (missing `anthropic` package in the dev venv, unrelated)

Closes #98406 (and substantively addresses the convergence arm of #72316).

### DIFF
--- a/agent/conversation_loop.py
+++ b/agent/conversation_loop.py
@@ -3611,6 +3611,14 @@ def run_conversation(
                     _finish_result = _cc_fr.normalize_response(response)
                     finish_reason = _finish_result.finish_reason
                     assistant_message = _finish_result
+                    # A clean slate for the per-API-call heuristic marker:
+                    # _should_treat_stop_as_truncated below stamps it True only
+                    # when IT rewrites stop→length, and the truncation handler
+                    # consumes it right after.  Clearing here (before the
+                    # evaluation, not after) guarantees no stale True from a
+                    # previous call can leak into this response's handler
+                    # branch (#98406).
+                    agent._suspicious_stop_rewrite = False
                     if agent._should_treat_stop_as_truncated(
                         finish_reason,
                         assistant_message,
@@ -4027,10 +4035,17 @@ def run_conversation(
                             # cap, so "still truncated" is wrong framing — the
                             # stitched partial IS the model's answer.  Complete
                             # the turn instead of reporting a truncation error
-                            # (#98406).  A genuine cap that follows later marks
-                            # itself with _suspicious_stop_rewrite=False and
-                            # runs the full 4-attempt loop as before.
-                            _converged_heuristic = not _should_request_continuation and _heuristic_truncation
+                            # (#98406).  Gated on a non-empty stitch: an empty
+                            # result is NOT a complete answer and belongs to the
+                            # empty-response recovery upstream.  A genuine cap
+                            # that follows later marks itself with
+                            # _suspicious_stop_rewrite=False and runs the full
+                            # 4-attempt loop as before.
+                            _converged_heuristic = (
+                                not _should_request_continuation
+                                and _heuristic_truncation
+                                and bool(truncated_response_parts)
+                            )
                             if partial_response and not _converged_heuristic:
                                 agent._vprint(
                                     f"{agent.log_prefix}⚠️  Response still truncated "
--- a/agent/conversation_loop.py
+++ b/agent/conversation_loop.py
@@ -3933,6 +3933,30 @@ def run_conversation(
                             )
                         if assistant_message is not None and not _trunc_has_tool_calls:
                             length_continue_retries += 1
+                            # ── Heuristic-guided stop rewrite converges fast (#98406) ──
+                            # When this truncation was manufactured by the
+                            # Ollama/GLM punctuation heuristic
+                            # (_should_treat_stop_as_truncated rewrote a
+                            # provider-correct ``stop`` to ``length``), one
+                            # continuation attempt is allowed (that preserves
+                            # the salvage behavior 8011aa31ba was written for),
+                            # but a second suspicious event converges instead
+                            # of burning the full retry budget: with a reasoning
+                            # model the injected nudge itself becomes the topic
+                            # of the next reasoning pass — extra full-budget
+                            # retries just produce another unpunctuated tail
+                            # and deepen the false truncation.
+                            _heuristic_truncation = getattr(
+                                agent, "_suspicious_stop_rewrite", False
+                            )
+                            agent._suspicious_stop_rewrite = False
+                            _should_request_continuation = (
+                                length_continue_retries < 4
+                                and not (
+                                    _heuristic_truncation
+                                    and length_continue_retries >= 2
+                                )
+                            )
                             # An EMPTY partial-stream stub (stream dropped
                             # mid tool-call before any text was delivered)
                             # must not be appended as an interim assistant
@@ -3957,7 +3981,7 @@ def run_conversation(
                                 if assistant_message.content:
                                     truncated_response_parts.append(assistant_message.content)
 
-                            if length_continue_retries < 4:
+                            if _should_request_continuation:
                                 _is_partial_stream_stub = (
                                     getattr(response, "id", "") == PARTIAL_STREAM_STUB_ID
                                 )
@@ -3999,11 +4023,26 @@ def run_conversation(
                                 break
 
                             partial_response = agent._strip_think_blocks(_join_truncated_parts(truncated_response_parts)).strip()
-                            if partial_response:
+                            # Heuristic-guided rewrites never saw a real output
+                            # cap, so "still truncated" is wrong framing — the
+                            # stitched partial IS the model's answer.  Complete
+                            # the turn instead of reporting a truncation error
+                            # (#98406).  A genuine cap that follows later marks
+                            # itself with _suspicious_stop_rewrite=False and
+                            # runs the full 4-attempt loop as before.
+                            _converged_heuristic = not _should_request_continuation and _heuristic_truncation
+                            if partial_response and not _converged_heuristic:
                                 agent._vprint(
                                     f"{agent.log_prefix}⚠️  Response still truncated "
                                     f"after 4 continuation attempts — keeping the "
                                     f"partial response received so far.",
+                                    force=True,
+                                )
+                            elif _converged_heuristic:
+                                agent._vprint(
+                                    f"{agent.log_prefix}ℹ️  Ollama/GLM stop heuristic "
+                                    f"fired once — accepting stitched response as "
+                                    f"complete.",
                                     force=True,
                                 )
                             # Unanswered continue nudges made every later turn re-truncate.
@@ -4036,9 +4075,13 @@ def run_conversation(
                                 "final_response": partial_response or None,
                                 "messages": messages,
                                 "api_calls": api_call_count,
-                                "completed": False,
-                                "partial": True,
-                                "error": "Response remained truncated after 4 continuation attempts",
+                                "completed": _converged_heuristic,
+                                "partial": not _converged_heuristic,
+                                "error": (
+                                    None
+                                    if _converged_heuristic
+                                    else "Response remained truncated after 4 continuation attempts"
+                                ),
                             }
 
                     if agent.api_mode in {"chat_completions", "bedrock_converse", "anthropic_messages"}:
--- a/run_agent.py
+++ b/run_agent.py
@@ -1795,6 +1795,19 @@ class AIAgent:
         (LiteLLM/sglang/vLLM/LM Studio proxies, Tailscale boxes), which
         report finish_reason correctly and were the source of #13971's
         false-positive truncation continuations.
+
+        Identity resolution notes:
+        - The ``glm`` substring check runs on the resolved model id, so
+          aliases that carry a GLM token (``glm-5.3-flash:cloud``,
+          ``zai/glm-5`` …) match, while a bare alias that hides the
+          model family entirely does not.
+        - The port/URL check runs on the resolved base URL: a custom
+          base URL still matches when it contains ``ollama`` or ``:11434``,
+          and a reverse proxy serving Ollama on another port matches
+          only if the provider is explicitly named ``ollama``.
+        - ``provider=="zai"`` routes are included regardless of URL
+          because Z.AI's GLM endpoints were the other half of the
+          original misreport family.
         """
         model_lower = (self.model or "").lower()
         provider_lower = (self.provider or "").lower()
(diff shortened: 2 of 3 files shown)


CryptoFamilyNFT:
Pushed a regression suite (256c77d1bd) that pins the fix boundaries:

**New tests (5):**
1. `test_suspicious_stop_flag_lifecycle` — the `_suspicious_stop_rewrite` marker is stamped only on a real heuristic rewrite; every other evaluation (non-GLM model, punctuated ending, genuine `length`) clears it
2. `test_pure_genuine_length_cap_still_gets_full_continuation_loop` — a genuine `finish_reason='length'` (never a stop rewrite) still gets the standard continuation nudge and loop
3. `test_ollama_glm_suspicious_stop_converges_after_one_continuation` — the live #98406 shape: tool turn → suspicious stop → suspicious stop after continuation → converge with `completed=True`, `error=None` (3 API calls, not 5)
4. `test_think_only_response_never_triggers_heuristic` — structural guarantee: reason-only content (<20 visible chars) can never fire the heuristic, so the convergent path can never mark an empty stitch as complete
5. `test_second_suspicious_event_converges_not_third_retry` — pins the third call as the last one

**Hardening added in the same push:**
- The flag is cleared in `conversation_loop` *before* `_should_treat_stop_as_truncated` evaluates each response (not after), so a stale `True` from a previous API call can never classify a genuine cap as heuristic-driven
- Convergence is additionally gated on `truncated_response_parts` being non-empty

**Full `tests/run_agent/` suite: 1922 passed, 7 failed — all 7 verified pre-existing on a clean main checkout via `git stash` (missing `anthropic` package in dev venv, unrelated to this change).**

liyangbing:
<p>This is a focused convergence fix for a provider-correct <code>stop</code> that the Ollama/GLM punctuation heuristic rewrites as <code>length</code>. Preserving one salvage continuation while accepting a non-empty stitched response on the second suspicious event avoids spending four full generations on a false truncation, while genuine provider <code>length</code> responses retain the existing ceiling.</p>

<h3>Suggested contract</h3>

<ul>
<li>Keep the heuristic marker strictly per API call and clear it before every finish-reason evaluation; a prior suspicious stop must never classify a later genuine <code>length</code> response.</li>
<li>Define the convergence invariant in terms of suspicious events, not only the global continuation counter: exactly one salvage continuation is allowed, then a non-empty stitched answer completes without the four-attempt error.</li>
<li>Never mark an empty or think-only stitch complete. Preserve the existing empty-response and tool-call recovery paths when the heuristic fires without visible answer text.</li>
<li>Keep genuine provider <code>length</code> behavior unchanged, including its continuation prompt, four-attempt ceiling, partial-response flag, and error wording.</li>
<li>Ensure the converged result is persisted and surfaced consistently in sync/async or streaming/non-streaming paths, with no leftover continuation nudge treated as a user turn on the next request.</li>
<li>Keep the heuristic narrow to the intended Ollama-hosted GLM route and document how model aliases, custom base URLs, and provider metadata affect that identity check.</li>
</ul>

<p>The new tests exercise the main loop and flag lifecycle. One additional provider-spy assertion would make the safety property explicit: after the second suspicious stop, no fourth API call is made, the final response is saved as complete, and no truncation error or <code>partial</code> marker is emitted. It would also be useful to test a partial first continuation followed by an empty second response.</p>

<h3>Regression matrix</h3>

<ul>
<li>tool turn followed by one suspicious stop and one suspicious continuation stop</li>
<li>two suspicious stops with a non-empty stitch, an empty stitch, and think-only content</li>
<li>a genuine provider <code>length</code> response before and after a suspicious-stop call</li>
<li>punctuated GLM output, short output, no whitespace, non-GLM model, and non-Ollama backend</li>
<li>streaming versus non-streaming responses, tool calls, partial stream stubs, and cancellation</li>
<li>custom model aliases/base URLs and route metadata preservation</li>
<li>persistence, retry/continuation counters, user-visible diagnostics, and absence of a fourth call after convergence</li>
</ul>

<p>Evidence reviewed: <code>agent/conversation_loop.py</code>, <code>run_agent.py</code>, and <code>tests/run_agent/test_run_agent.py</code> in PR head <code>256c77d1bdf54e339c4c3cfc3bf4e0920dd5499c</code>.</p>

CryptoFamilyNFT:
Thanks for the careful review — the contract summary matches what we intended, and most of it was already covered by the first two commits (`35d7c3f`, `256c77d`):

- per-API-call marker cleared **before** each finish-reason evaluation → done in `256c77d`
- convergence defined on suspicious events, not the global counter → that's exactly how `_should_request_continuation` gates (`_heuristic_truncation and length_continue_retries >= 2`)
- empty/think-only stitches can never be marked complete → gated on `truncated_response_parts` being non-empty, plus the predicate-level structural guard (think-only content can't fire the heuristic at all — `test_think_only_response_never_triggers_heuristic`)
- genuine `length` behavior byte-identical (prompt, ceiling, partial flag, error wording) → pinned by `test_pure_genuine_length_cap_still_gets_full_continuation_loop`

For the two gaps you flagged, pushed `3cd75f27`:

**Provider-spy assertions** — `test_second_suspicious_event_converges_not_third_retry` now additionally asserts `create.call_count == 3` (no fourth call after the second suspicious stop), that the persisted transcript ends with the stitched assistant answer, and that no `_length_continuation_nudge` survives as a pending user turn for the next request.

**Partial-then-empty scenario** — new `test_suspicious_continuation_followed_by_empty_response`: suspicious stop → partial stitched, salvage continuation returns empty. The convergent path refuses to complete on the empty stitch; the loop recovers through the normal path and the final answer joins whatever the recovery produced.

**Identity docs** — also expanded the `_is_ollama_glm_backend` docstring with how model aliases, custom base URLs, and provider metadata interact with the `glm`/`ollama`/`:11434` checks, per your contract note on keeping the heuristic narrow.

One point we did not touch: the sync/streaming surfacing audit (contract item 5). The convergent exit reuses the exact same return dict shape and `_persist_session` call as the pre-existing 4-attempt exit, so streaming consumers see the same fields — but we haven't exercised the SSE path specifically. Happy to add that if you want it in this PR rather than a follow-up.

Full suite after the additions: 286 passed on `test_run_agent.py` (the single `TestAnthropicInterruptHandler` failure is the pre-existing missing-`anthropic` env issue, verified identical on a clean checkout).

0xSteveTheHead:
@teknium1 bumping this for review when you get a chance.
The PR addresses the Ollama/GLM false-truncation loop while preserving the existing salvage behavior and keeping genuine length continuations unchanged. Added regression coverage for the failure mode as well.
Would appreciate your eyes on itt

teknium1:
Closing with credit; the root cause you identified (#98406) is fixed on main via https://github.com/NousResearch/hermes-agent/pull/100924. We went with the identity-gate approach instead: Ollama Cloud (`ollama.com` host or `:cloud` model suffix) is now exempt from `_should_treat_stop_as_truncated` entirely, so a provider-correct `stop` is never rewritten and no continuation nudge is sent. With cloud exempt, the one-attempt convergence cap only applies to local Ollama, where the original 8011aa31ba salvage behavior is intentionally retained, so we kept the smaller change. Thanks for the thorough investigation and the regression-shape write-up; it drove the live repro.

Issue #97975, opened 2026-08-29
Builtin input() prompts leak raw modifyOtherKeys escapes — separate code path from the prompt_toolkit Shift+letter fixes


MKappes:
### Environment

- Terminal: Ghostty 1.3.1 (macOS 26.6.2), `TERM=xterm-ghostty`, `TERM_PROGRAM=ghostty`
- Hermes Agent v0.20.6 (2026.8.27)
- Affects: every `input()` call site reached while modifyOtherKeys is active

### Summary

`cli.py` pushes xterm modifyOtherKeys level 2 (`CSI >4;2m`) for allowlisted terminals (`_enable_extended_enter_keys`, cli.py:4281; Ghostty gets modifyOtherKeys only, cli.py:4314). While that mode is active the terminal re-encodes every Shift combo as `ESC[27;<mod>;<codepoint>~`.

The prompt_toolkit input path decodes those via `ANSI_SEQUENCES` aliases. **Python's builtin `input()` does not.** Any prompt that uses `input()` instead of prompt_toolkit receives the raw escape text.

This is a separate code path from #94921 / #92343 / #95550 / #97413, which all concern the prompt_toolkit parser. No alias table or `KeyPress.data` fix reaches `input()`, so those fixes (including PR #97503, which I authored) leave this surface broken.

### Reproduction — standalone, no Hermes config required

```python
#!/usr/bin/env python3
import sys
sys.stdout.write("\x1b[>4;2m")   # exactly what cli.py:4314 sends to Ghostty
sys.stdout.flush()
try:
    answer = input("Type THIS (Shift+T,H,I,S) then Enter: ")
finally:
    sys.stdout.write("\x1b[>4;0m")
    sys.stdout.flush()
print("repr(received) =", repr(answer))
```

Under Ghostty, pressing Shift+T, Shift+H, Shift+I, Shift+S:

```
Type THIS (Shift+T,H,I,S) then Enter: ^[[27;2;84~^[[27;2;72~^[[27;2;73~^[[27;2;83~
repr(received) = '\x1b[27;2;84~\x1b[27;2;72~\x1b[27;2;73~\x1b[27;2;83~'
RESULT: LEAK — raw escape sequences reached the application
```

Verified under tmux with `TERM=xterm-ghostty`, feeding the raw bytes the terminal actually emits.

### Where I hit it in real use

`agent/shell_hooks.py:969`, the shell-hook consent gate:

```
[...]

PR #98413, opened 2026-08-30
fix(cli): decode modifyOtherKeys/Kitty escapes leaked into bare input() prompts

lEWFkRAD:
## What

The classic CLI enables extended-key reporting (`modifyOtherKeys` level 2, and Kitty keyboard protocol on allowlisted terminals) globally for the process so Shift+Enter works in the prompt-toolkit chat. prompt_toolkit decodes the resulting sequences — but Python's builtin `input()` does not. Every bare `input()` prompt (hook consent, credential import, MCP setup, profile destruction, the hindsight wizard, the ACP browser-tools install) receives raw escape text like `^[[27;2;84~` instead of `T` whenever the user types a shifted character (#97975).

## Root cause

`cli.py::_enable_extended_enter_keys` pushes `CSI >4;2m` / `CSI >1u` for the whole process lifetime. The chat REPL reads through prompt_toolkit (which decodes via `ANSI_SEQUENCES`, extended by `hermes_cli.pt_input_extras`), but setup/consent code paths use builtin `input()`, which has no decoder — so the escape bytes land in the answer verbatim. Free-text prompts (API keys, endpoints) silently store corrupted values.

## Fix

New `hermes_cli/input_decode.py`:

- `decode_modified_key_sequences()` — a pure string transform. Decodes `ESC[27;2;<cp>~` and Kitty `ESC[<cp>;2u` (plus Kitty alternate-report mods 3/4) to `chr(cp)`; the codepoint is already layout-applied by the terminal, so shift-only decoding is just `chr`. Everything else passes through byte-for-byte: ctrl/alt/super combos, Alt under modifyOtherKeys (mod 3 is ambiguous there, never decoded), functional keys (Kitty private-use plane incl. Shift+Enter 57427), control codepoints, lock-bit combos (CapsLock is applied to the codepoint before encoding — decoding mod 65 would rewrite an *un-shifted* keypress), and out-of-range/surrogate codepoints.
- `safe_input()` — drop-in `input()` wrapper.

Migrated all bare `input()` prompt sites named in the issue: `agent/shell_hooks.py` (consent gate), `hermes_cli/auth_commands.py` ×4, `hermes_cli/mcp_config.py` ×2, `hermes_cli/profiles.py` (destruction confirmation), `acp_adapter/entry.py`, `plugins/memory/hindsight/__init__.py` ×4. `line_input()` sites already route through prompt_toolkit and were left alone. Remaining bare `input()` calls elsewhere are non-interactive/error-path reads; the wrapper is exported for those to adopt incrementally.

## Test plan

`tests/hermes_cli/test_input_decode.py` — 15 deterministic tests, no TTY: the issue's exact repro (`ESC[27;2;84~` → `T`), mixed `"Type THIS"` round-trip, Kitty CSI-u incl. alternate-report mods, ctrl/alt/super passthrough, Alt-under-MOK passthrough, lock-bit passthrough, functional-key/control/surrogate/out-of-range passthrough, unrelated escapes (bracketed paste, SGR, cursor) untouched, `safe_input` decode + EOF propagation.

Real pytest counts:
- new tests: 15 passed
- regression (`test_input_sanitize.py`, `test_cli_output.py`): all passed
- `tests/agent/test_shell_hooks_consent.py`: 1 failure (`test_tilde_path_approval_records_resolvable_mtime`) **reproduces identically on clean origin/main** (HOME-tilde expansion in the approval store on this Windows runner) — pre-existing, unrelated to this diff.

Contributes to #97975 (not "Closes" — the ~60-site full sweep is intentionally incremental; the named high-risk sites are done).

### DIFF
--- a/acp_adapter/entry.py
+++ b/acp_adapter/entry.py
@@ -179,7 +179,9 @@ def _run_setup() -> None:
     if not sys.stdin.isatty():
         return
     try:
-        reply = input(
+        from hermes_cli.input_decode import safe_input
+
+        reply = safe_input(
             "\nInstall browser tools? Downloads agent-browser (npm) and "
             "optionally Playwright Chromium (~400 MB). [y/N] "
         ).strip().lower()
--- a/agent/shell_hooks.py
+++ b/agent/shell_hooks.py
@@ -966,7 +966,12 @@ def _prompt_and_record(
         f"  commands you trust."
     )
     try:
-        answer = input("Allow this hook to run? [y/N]: ").strip().lower()
+        # safe_input, not bare input(): the classic CLI pushes
+        # modifyOtherKeys/Kitty extended-key mode globally, and a shifted
+        # answer would otherwise arrive as raw escape text (#97975).
+        from hermes_cli.input_decode import safe_input
+
+        answer = safe_input("Allow this hook to run? [y/N]: ").strip().lower()
     except (EOFError, KeyboardInterrupt):
         print()  # keep the terminal tidy after ^C
         return False
--- a/hermes_cli/auth_commands.py
+++ b/hermes_cli/auth_commands.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 from hermes_cli.cli_output import line_input
+from hermes_cli.input_decode import safe_input
 
 import math
 import sys
@@ -353,7 +354,7 @@ def auth_add_command(args) -> None:
             else:
                 print("Found existing shared Nous OAuth credentials")
             try:
-                do_import = input("Import these credentials? [Y/n]: ").strip().lower()
+                do_import = safe_input("Import these credentials? [Y/n]: ").strip().lower()
             except (EOFError, KeyboardInterrupt):
                 do_import = "y"
             if do_import in {"", "y", "yes"}:
@@ -733,7 +734,7 @@ def _interactive_auth() -> None:
         print(f"  {i}. {choice}")
 
     try:
-        raw = input("\nChoice: ").strip()
+        raw = safe_input("\nChoice: ").strip()
     except (EOFError, KeyboardInterrupt):
         return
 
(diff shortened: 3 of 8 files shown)


Enough1122:

**Overall:** Decodes Kitty/modifyOtherKeys escapes leaked into bare input() prompts.

**What it does**
- New `hermes_cli/input_decode.py` decodes CSI-u / modifyOtherKeys sequences; replaces 7 bare `input()` calls with `safe_input()` across `acp_adapter/entry.py`, `agent/shell_hooks.py`, `hermes_cli/auth_commands.py` etc. Classic CLI had enabled extended-key mode globally for prompt_toolkit but bare input() got raw escape text.
- Keeps EOF/KeyboardInterrupt handling intact.

**Non-blocking notes**
- New module should handle plain input passthrough unchanged (no double-decode) and be import-safe early in startup — appears lazy-imported per call site, good.

No blocking issues found.

andrexibiza (commented):
Blocking finding on exact head `2740339a6e8a92285471521f2e15318fcd521a1c`: the Kitty CSI-u modifier semantics in `input_decode.py` are wrong, and the tests currently encode the same mistake.

`_SHIFT_ONLY_MODS = {2, 3, 4}` treats CSI-u modifier 3/4 as repeat/release forms of Shift and rewrites `ESC[84;3u` / `ESC[84;4u` to literal `T`. In Hermes' existing keyboard owner, `hermes_cli/pt_input_extras.py`, those values are already canonicalized as ordinary modifier masks: 2 = Shift, 3 = Alt, 4 = Shift+Alt, 5 = Ctrl, 6 = Ctrl+Shift, etc. Event type is a separate Kitty field; it is not represented by incrementing the modifier value. So this decoder currently strips Alt (and Shift+Alt) from input at credential/consent/setup prompts, creating a new silent input-corruption path while fixing #97975.

Please restrict the plain `CSI <codepoint>;<modifier> u` decoder to the actual Shift-only modifier (`2`) unless the parser is extended to consume Kitty's separate event-type syntax correctly. Flip the current `;3u` / `;4u` expectations into preservation regressions, and add any repeat/release coverage using the protocol's real event field rather than modifier substitution.

The one-commit head is otherwise exact-head green in CI `33297988692`, Docker `33297988313`, and Nix `33297988346`; this is a source-semantics blocker, not a CI one.

PR #78202, opened 2026-08-04
fix(tts): piper lazy-install parity + probe install-safety + null-config guards

andrexibiza:
Related #44001 #44010 #47318 #61799 #77040 #77932
## What changed and why

This PR completes the **provider registry & lazy-install consistency** bug class: built-in providers must be lazy-installable on fresh installs, requirements probes must never trigger pip installs, and provider fallback/readiness reporting must be truthful. It salvages and completes two open PRs (authorship preserved, see credits) plus the remaining null-config gap.

### 1. Piper TTS is now lazy-installable (issue #44001)

- `tools/lazy_deps.py`: added the missing `tts.piper` feature entry (was absent — fresh installs with `tts.provider: piper` failed with a manual-install error while edge/elevenlabs/mistral auto-installed).
- `tools/tts_tool.py::_import_piper`: calls `lazy_deps.ensure("tts.piper")` before importing, matching the Edge/ElevenLabs/Mistral pattern.
- **Pin parity** (maintainer-flagged gap in #44010): `piper-tts==1.4.2` is now pinned in all three places — the new `[piper]` extra in `pyproject.toml`, the `tts.piper` lazy entry, and the `hermes tools` post-setup installer (previously unpinned `-U piper-tts`, which could install a newer version and then get **downgraded** by the lazy pin on first use). `uv.lock` regenerated (adds `piper-tts` + `pathvalidate`).
- `tests/test_project_metadata.py`: `[piper]` added to the lazy-covered-extras contract so it can't leak into `[all]`.
- `tests/tools/test_piper_lazy_install.py`: environment-independent (mocks `find_spec` instead of assuming piper is absent) and asserts the lazy pin matches the pyproject extra.

### 2. Requirements probes never lazy-install (follow-up to #77040)

- `tools/transcription_tools.py::_get_provider`: new `install: bool = True` kwarg; probe callers pass `install=False` so a bare status check can never freeze on a pip install.
- `tools/voice_mode.py::check_voice_requirements` and `tools/wake_word.py::_stt_ready` call it with `install=False`.
- **Activation gate preserved** (maintainer-flagged gap in #77040): an explicit uninstalled `local` STT provider now counts as *ready* when lazy installs are permitted — so classic `/voice on` still activates and the install happens at first real transcription, instead of rejecting an installable provider before the transcription path could install it. The probe itself never runs pip (regression test asserts `install=False` is passed).
[...]

PR #77040, opened 2026-08-02
fix(stt): keep requirements probes from lazy-installing faster-whisper

Limkyy:
## Bug Description

Starting `hermes --tui` on a machine with `stt.provider: local` and
faster-whisper not yet installed hangs on the `forging session…` screen
for minutes (or until killed).

The TUI startup handshake (`wake.start`) calls
`check_wake_word_requirements()` → `_stt_ready()` → `_get_provider()`, and
`_get_provider` resolves an uninstalled explicit `local` provider by
running a **synchronous lazy pip install**
(`lazy_deps.ensure("stt.faster_whisper")` → `uv pip install
faster-whisper`, up to 300s for the ~200MB ctranslate2 wheel). The
handshake is handled on the gateway's main thread, so the Ink frontend
never receives a ready event and sits on `forging session…` until the
install finishes, times out, or the process is killed.

`check_voice_requirements()` (used by the TUI's `/voice status` RPC) has
the same freeze.

## Root Cause

`tools/transcription_tools.py::_get_provider` calls
`_try_lazy_install_stt()` from both the explicit-`local` branch and the
auto-detect branch. Any caller that merely wants to *probe* STT readiness
(resolve which provider would be used) therefore triggers a real,
synchronous, multi-minute pip install.

The TTS path already solved this exact problem:
`_tts_ready()` in `tools/wake_word.py` is documented as
"PROBE, not an installer" — it never touches pip from a status poll, and
treats uninstalled-but-lazy-installable providers as ready (installed at
first real use). STT never got the same treatment.

## Fix

- `_get_provider(stt_config, *, install=True)` — new keyword-only
  `install` switch (default `True`, fully backward compatible). When
  `False`, an uninstalled `local` provider is *not* pip-installed; the
  resolver just falls through to `local_command` / cloud providers /
  `"none"`.
- `_stt_ready()` now resolves with `install=False` and, mirroring
  `_tts_ready`, treats an explicit `local`/`local_command` provider that
  isn't installed yet as ready **iff** lazy installs are permitted — the
  actual fetch happens lazily at first real transcription, never from a
  startup handshake or status poll.
- `check_voice_requirements()` resolves with `install=False` too (same
  freeze via the TUI `/voice status` RPC).

## How to Verify

1. Set `stt.enabled: true` / `stt.provider: local` with faster-whisper
   absent from the venv.
2. `hermes --tui` — the UI should reach the main screen immediately
   instead of hanging on `forging session…`.
[...]

PR #86155, opened 2026-08-14
fix(tui): skip wake checks while wake word is disabled

BlackishGreen33:
## Related Issue

Fixes #86131

## What does this PR do?

The TUI passively calls `wake.start` when the gateway becomes ready. When wake word was disabled, the handler still resolved the capture mode and checked optional voice dependencies before returning `disabled`. Those checks can start a lazy STT install even though no listener should be armed.

This change applies the config and surface gate first for passive requests. An explicit `persist: true` enable action still checks requirements before saving `wake_word.enabled`, so a setup that cannot arm is not recorded as enabled.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✅ Tests (adding or improving test coverage)

## Changes Made

- Return the existing `disabled` or `disabled_for_surface` result before capture and dependency checks when the request is not enabling wake word.
- Keep the requirements-first behavior for an explicit enable action.
- Extend the existing toggle test and add a temporary-`HERMES_HOME` config-path regression test.

## How to Test

1. Run `./venv/bin/python -m pytest tests/test_tui_gateway_server.py -k wake -q` (`4 passed`).
2. Run `./venv/bin/python -m pytest tests/test_tui_gateway_server.py -q` (`553 passed`).
3. Run `uvx --from ruff==0.15.10 ruff check tui_gateway/server.py tests/test_tui_gateway_server.py` (`All checks passed`).
4. Run `git diff --check origin/main...HEAD` (no errors).

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
- [x] My PR contains **only** changes related to this fix/feature (no unrelated commits)
- [ ] I've run `pytest tests/ -q` and all tests pass
- [x] I've added tests for my changes (required for bug fixes, strongly encouraged for features)
- [x] I've tested on my platform: macOS 26.5.1

### Documentation & Housekeeping

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A, behavior and configuration are unchanged
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — N/A, no config keys changed
[...]

PR #98414, opened 2026-08-30
fix(tui_gateway): refuse disabled wake-word auto-arm before the requirements probe

### LINKED ISSUE #98409: [Bug]: TUI shows spurious 'error: timeout: wake.start' when wake word disabled (STT lazy-install blocks RPC)
## Bug Description

On a machine with no STT provider configured (wake word disabled in config), the TUI shows a spurious `error: timeout: wake.start` at the bottom of every chat.

The TUI auto-arms the "Hey Hermes" listener on every `gateway.ready` by firing a fire-and-forget `wake.start` RPC (`ui-tui/src/app/createGatewayEventHandler.ts:669`). Even though `wake_word.enabled` is `False`, the server runs the full requirement check **before** honoring the disabled state:

```
wake.start → check_wake_word_requirements() → _stt_ready()
  → _get_provider() → _try_lazy_install_stt()
  → lazy_deps.ensure("stt.faster_whisper")
  → synchronous subprocess.run(uv pip install faster-whisper)   ← blocks
```

Installing faster-whisper (a ~200MB ctranslate2 wheel) takes far longer than the client RPC timeout (`REQUEST_TIMEOUT_MS`, default 120s), so the client rejects with `Error: timeout: wake.start` (`ui-tui/src/gatewayClient.ts:761`). The crash log confirms the pip install is the thing timing out.

## Steps to Reproduce

1. Fresh install, no STT provider configured (wake word disabled by default).
2. Launch `hermes --tui`.
3. Observe `error: timeout: wake.start` at the bottom of the chat.

## Expected Behavior

- With `wake_word.enabled: False`, the server should short-circuit and return `{started: False, reason: "disabled"}` **before** running the expensive STT/install probe.
- The auto-arm path should never trigger a synchronous pip install at all.

## Actual Behavior

The disabled wake word still triggers a synchronous `faster-whisper` pip install inside the RPC handler thread, which exceeds the client timeout and surfaces a confusing `error: timeout: wake.start` on every connect.

## Related

- #77040 `fix(stt): keep requirements probes from lazy-installing faster-whisper` — this is the root-cause fix (don't lazy-install inside probes), but it was **closed unmerged**.
- #85438 `fix(gateway): pool-route voice + wake RPCs so STT lazy install can't block WS sends` — merged, but only moves the RPC off the WS reader thread; the handler thread still blocks on the install, so the client-side RPC timeout still fires.

## Suggested Fix

1. In `tui_gateway/server.py` `wake.start`, check `wake_word.enabled` (and surface scoping) **before** calling `check_wake_word_requirements()`, so a disabled wake word returns immediately without probing STT/TTS.
2. Ensure `_get_provider()` / `_stt_ready()` never trigger a synchronous lazy install from a probe path (resurrect the intent of #77040).

## Environment

- OS: Android 16 (Termux)
- Surface: TUI
- STT: none configured

liuhao1024:
## What does this PR do?

`wake.start` ran `check_wake_word_requirements()` **before** the enabled/surface gate. On a machine with no STT provider configured (wake word disabled in config), that probe walks `check_wake_word_requirements() → _stt_ready() → _get_provider() → _try_lazy_install_stt() → lazy_deps.ensure("stt.faster_whisper")` — a synchronous `uv/pip` subprocess installing a ~200MB ctranslate2 wheel. The TUI auto-arms the "Hey Hermes" listener on every `gateway.ready` by firing a fire-and-forget `wake.start`, so on such machines the install far outlives the client RPC timeout (`REQUEST_TIMEOUT_MS`, default 120s) and every chat shows a spurious `error: timeout: wake.start`.

An auto-arm (no `persist`) against a wake word that cannot arm is refused with `reason: "disabled"` / `"disabled_for_surface"` regardless of what the probe would say — so the refusal now happens **before** the probe, which is the lazy-install entry point. The `persist: true` gesture path is unchanged and keeps requirements-first, so an explicit gesture on an unarmed-able setup (no STT/TTS, no mic, missing key) still refuses **without** flipping `wake_word.enabled`.

This is the server-side half of #98409's suggested fix (item 1). Item 2 (probes never lazy-installing at all, the #77040 intent) is being carried by open #78202 at the `tools/` layer; the two changes are orthogonal file-wise and complementary behavior-wise.

## Related Issue

Fixes #98409

## Type of Change

## Changes Made

- `tui_gateway/server.py` — in the `wake.start` handler, refuse non-`persist` calls with `{started: False, reason: "disabled" | "disabled_for_surface"}` before `check_wake_word_requirements()` runs. Same reasons, same log shape as the previous post-probe refusal; only the ordering changes.
- `tests/test_tui_gateway_server.py` — new regression test `test_wake_start_auto_arm_disabled_skips_requirements_probe`: monkeypatches `check_wake_word_requirements` to raise if called, then asserts both the `disabled` and the `disabled_for_surface` auto-arm paths return the refusal without ever reaching the probe.

## How to Test

1. `pytest tests/test_tui_gateway_server.py tests/tui_gateway/test_protocol.py -q` — Observed result: 686 passed (includes the new test; the pre-existing `test_wake_toggle_persists_enabled_flag_only_on_explicit_gesture` still passes, confirming the `persist` gesture keeps requirements-first semantics).
2. `ruff check tui_gateway/server.py tests/test_tui_gateway_server.py` — Observed result: all checks passed.
3. Manual equivalent of the issue repro (fresh config, `wake_word.enabled: false`, no STT): `wake.start` auto-arm should now return `{started: False, reason: "disabled"}` immediately, with no `faster-whisper` install attempt and no `error: timeout: wake.start` in the TUI.

### Code

- [x] N/A — behavior-preserving reorder inside one handler; the handler docstring's gate description already covers both refusal reasons
- [x] N/A — no config keys changed
- [x] N/A — no architecture or workflow changes
- [x] N/A — pure server-side ordering change, no platform-specific I/O touched
- [x] N/A — no tool behavior changes

### DIFF
--- a/tests/test_tui_gateway_server.py
+++ b/tests/test_tui_gateway_server.py
@@ -2112,6 +2112,58 @@ def test_wake_owner_is_sticky_and_routes_detection_to_first_transport(monkeypatc
         server._wake_owner_surface = ""
 
 
+def test_wake_start_auto_arm_disabled_skips_requirements_probe(monkeypatch):
+    """Auto-arm on a wake word that cannot arm must refuse BEFORE probing.
+
+    check_wake_word_requirements() → _stt_ready() can lazy-install
+    faster-whisper as a synchronous subprocess that outlives the client's RPC
+    timeout, so a disabled-config auto-arm on every gateway.ready used to
+    surface as a spurious "error: timeout: wake.start" (#98409). The refusal
+    is reason:"disabled"/"disabled_for_surface" regardless of what the probe
+    would say, so the probe (the lazy-install entry point) must not run at
+    all on this path.
+    """
+    from tools import wake_word
+
+    config = {"enabled": False, "phrase": "hey hermes", "surface": "auto",
+              "start_new_session": True}
+    monkeypatch.setattr(wake_word, "load_wake_word_config", lambda: dict(config))
+
+    def _probe_explodes(_cfg):
+        raise AssertionError(
+            "check_wake_word_requirements() ran for an auto-arm that cannot "
+            "arm — the STT lazy-install entry point must be unreachable here"
+        )
+
+    monkeypatch.setattr(wake_word, "check_wake_word_requirements", _probe_explodes)
+
+    transport = types.SimpleNamespace(_closed=False)
+    server._wake_owner_transport = None
+    server._wake_owner_surface = ""
+    try:
+        # Feature off in config.
+        disabled = _dispatch_sync({
+            "id": "wake-autoarm-disabled",
+            "method": "wake.start",
+            "params": {"surface": "gui"},
+        }, transport=transport)
+        assert disabled["result"] == {"started": False, "reason": "disabled"}
+
+        # Feature on but scoped to a different surface: same must-refuse
+        # conclusion, so the probe must not run here either.
+        config["enabled"] = True
+        config["surface"] = "tui"
+        scoped = _dispatch_sync({
+            "id": "wake-autoarm-scoped",
+            "method": "wake.start",
+            "params": {"surface": "gui"},
+        }, transport=transport)
+        assert scoped["result"] == {"started": False, "reason": "disabled_for_surface"}
+    finally:
+        server._wake_owner_transport = None
+        server._wake_owner_surface = ""
+
+
 def test_wake_toggle_persists_enabled_flag_only_on_explicit_gesture(monkeypatch):
     """The ear toggle / /wake on|off write wake_word.enabled; auto-arm never does."""
     from tools import wake_word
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #86155: both short-circuit passive disabled wake.start before requirements probing. #86155 additionally avoids capture-mode work and covers config-file loading.

liuhao1024:
You're right — this is a duplicate of #86155, and theirs is the stronger version: the gate sits before `resolve_capture_mode` (skipping capture work as well) and it adds the config-file loading regression test. My PR only moved the refusal after capture resolution, so it is a strict subset on the same `wake.start` path.

For the record, #98409 (which this PR targeted) reports the same root cause from a different angle: the passive auto-arm on a disabled wake word still ran `check_wake_word_requirements` → `_stt_ready`, whose lazy faster-whisper install outlives the client's RPC timeout and surfaces as `error: timeout: wake.start` on every `gateway.ready`. Closing this in favor of #86155 — the timeout scenario in #98409 may be worth adding there as an additional validation case.

PR #98418, opened 2026-08-30
fix(dashboard-auth): warn (not debug) when Portal client_id is unset (#98338)

Halldrix:
## What does this PR do?

Raises the log level from DEBUG to WARNING when the Nous Portal OAuth `client_id` is not configured, so the silently-disappearing Portal login is surfaced to operators.

### Problem

In `plugins/dashboard_auth/nous/__init__.py`, when no `client_id` is configured (neither `HERMES_DASHBOARD_OAUTH_CLIENT_ID` nor `dashboard.oauth.client_id` in `config.yaml`), `register()` set `LAST_SKIP_REASON` and logged it at DEBUG before returning. A hosted instance with the Portal fogged off therefore showed no operator-visible log at all — the gate silently skipped, with the skip reason only visible at DEBUG. The adjacent branch (client_id set but not matching the `agent:{instance_id}` contract) already logs WARNING, so this branch was the inconsistent outlier.

## Related Issue

**Refs #98338** (partial — satisfies the issue's **minor: `plugins/dashboard_auth/nous/__init__.py` logs `LAST_SKIP_REASON` at debug when no `client_id` is configured; should be warning**).

The other requests in #98338 (dual-logging of the refresh path, backoff/circuit breaker, device id on audit records, scheduler `max_parallel_jobs` logging, dashboard-auth.log rotation) are NOT addressed here and are tracked in separate PRs, so #98338 should stay open for them.

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `plugins/dashboard_auth/nous/__init__.py`: changed `logger.debug("dashboard-auth-nous: %s", LAST_SKIP_REASON)` to `logger.warning(...)` in the no-client-id branch only; the malformed-client-id branch was already WARNING and is untouched.
- `tests/plugins/dashboard_auth/test_nous_provider.py`: added `test_missing_client_id_logs_warning` asserting the skip is logged at WARNING (not DEBUG) when `client_id` is missing.
- `tests/agent/test_system_prompt_restore.py`: scoped two happy-path "no warning" asserts to the `agent.conversation_loop` logger (`r.name == "agent.conversation_loop"`). They previously scanned the global `caplog.records` list, which fails on ambient WARNINGs from unrelated plugins (dashboard-auth config) in some dev/CI environments — including the one this PR's new WARNING surfaced. The other two WARNING scans in the file filter by message and are unaffected.

## How to Test

1. `pytest tests/plugins/dashboard_auth/test_nous_provider.py -k client_id` → 6 passed (incl. the new test)
2. `pytest tests/agent/test_system_prompt_restore.py` → 16 passed (scoped asserts green)

### Code

- [x] I've tested on my platform: Debian GNU/Linux 13 (trixie), x86_64

## Screenshots / Logs

N/A

---

**🛠️ Dev:** Halldrix
**🤖 Sidekick:** Hermes Agent v0.20.5 — main `26350357d7`
**🐞 Reproduction:** ✅ Confirmed (tests green)

### DIFF
--- a/plugins/dashboard_auth/nous/__init__.py
+++ b/plugins/dashboard_auth/nous/__init__.py
@@ -130,12 +130,12 @@ def _settings() -> dict:
     portal_url = resolve_env_or_cfg("HERMES_DASHBOARD_PORTAL_URL", section.get("portal_url", "")) or _DEFAULT_PORTAL_URL
     if not client_id:
         raise SkipRegistration(
-            "HERMES_DASHBOARD_OAUTH_CLIENT_ID is not set (and dashboard.oauth.client_id "
-            "in config.yaml is empty). The Nous Portal provisions this env var (shape "
-            "'agent:{instance_id}') when it deploys a Hermes Agent instance — set it to "
-            "your provisioned client id (either as an env var or under "
-            "dashboard.oauth.client_id in config.yaml), or pass --insecure to skip the "
-            "OAuth gate entirely.")
+        "HERMES_DASHBOARD_OAUTH_CLIENT_ID is not set (and dashboard.oauth.client_id "
+        "in config.yaml is empty). The Nous Portal provisions this env var (shape "
+        "'agent:{instance_id}') when it deploys a Hermes Agent instance — set it to "
+        "your provisioned client id (either as an env var or under "
+        "dashboard.oauth.client_id in config.yaml), or pass --insecure to skip the "
+        "OAuth gate entirely.", level="warning")
     if not client_id.startswith("agent:"):
         raise SkipRegistration(
             f"HERMES_DASHBOARD_OAUTH_CLIENT_ID={client_id!r} doesn't match the contract "
@@ -165,6 +165,7 @@ import httpx  # noqa: F401,E402
 import os  # noqa: F401,E402
 import secrets  # noqa: F401,E402
 import urllib.parse  # noqa: F401,E402
+
 
 
 _PLUGIN_COMPAT_LAZY = {
--- a/tests/agent/test_system_prompt_restore.py
+++ b/tests/agent/test_system_prompt_restore.py
@@ -59,8 +59,14 @@ class TestStoredPromptReuse:
         assert agent._cached_system_prompt == stored
         agent._build_system_prompt.assert_not_called()
         db.update_system_prompt.assert_not_called()
-        # No warnings on the happy path
-        assert not [r for r in caplog.records if r.levelno >= logging.WARNING]
+        # No warnings on the happy path — scoped to the logger under test:
+        # unrelated plugins (dashboard-auth config warnings, etc.) emit their
+        # own WARNINGs in some dev/CI environments and are not this test's
+        # concern. The at_level() above already targets agent.conversation_loop.
+        assert not [
+            r for r in caplog.records
+            if r.name == "agent.conversation_loop" and r.levelno >= logging.WARNING
+        ]
 
     def test_present_row_with_unicode_preserved(self):
         """Non-ASCII bytes in the stored prompt are not mangled."""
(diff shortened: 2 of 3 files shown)


liyangbing:
<p>This fixes a real observability gap: a missing Portal <code>client_id</code> caused the provider to skip silently at the default log level, while the adjacent malformed-ID path already warned. Scoping the happy-path assertions to the logger under test also makes the regression test robust against unrelated plugin warnings.</p>

<h3>Suggested contract</h3>

<ul>
<li>Distinguish an intentional Portal disable from an accidental missing configuration. A warning is useful for the latter, but an explicit opt-out should not create recurring startup noise or imply that authentication is broken.</li>
<li>Keep configuration precedence observable without exposing credentials: identify whether the environment variable or YAML key was absent, invalid, or intentionally disabled, while never logging the client ID or OAuth material.</li>
<li>Make the warning actionable and stable. Include a bounded reason/code and the safe remediation location, and ensure repeated <code>register()</code> calls or gateway reloads do not flood logs with the same unchanged skip.</li>
<li>Reset skip state on configuration transitions. Removing an ID should warn and unregister the provider; adding a valid ID later should clear stale skip state and register normally.</li>
<li>Preserve fail-closed behavior: warning visibility must not cause partial provider registration, a misleading login button, or an OAuth request with an empty client ID.</li>
<li>Keep logger scoping consistent across CLI, gateway, hosted Portal, and desktop startup so operators can reliably route this warning without relying on global <code>caplog</code>-style assumptions.</li>
</ul>

<p>The added test proves the missing-ID record is WARNING rather than DEBUG. A useful follow-up is a transition test that calls <code>register()</code> with no ID, then a valid ID, then no ID again, asserting registration state, warning cardinality, and redaction remain correct; also cover an explicit disabled setting if the configuration model supports one.</p>

<h3>Regression matrix</h3>

<ul>
<li>missing environment ID and missing YAML ID, with each precedence combination</li>
<li>explicit disable, valid <code>agent:</code> ID, malformed ID, empty ID, and whitespace-only ID</li>
<li>first registration, repeated registration, gateway reload, and configuration transition</li>
<li>CLI, gateway, desktop, and hosted Portal startup log routing</li>
<li>warning versus debug/info severity, actionable reason text, and no client ID/token leakage</li>
<li>provider registration/unregistration state, login UI state, and no OAuth call on a missing ID</li>
<li>unrelated plugin warnings proving tests remain logger-scoped rather than globally fragile</li>
</ul>

<p>Evidence reviewed: <code>plugins/dashboard_auth/nous/__init__.py</code>, <code>tests/plugins/dashboard_auth/test_nous_provider.py</code>, and <code>tests/agent/test_system_prompt_restore.py</code> in PR head <code>0cfa85f859daafb3794fa3b667c72d9b58b2cf17</code>.</p>

Halldrix:
Thanks for the detailed review, @liyangbing — and for confirming the scoping fix on the flaky test, that's exactly what it was for.

Point-by-point against current head (`466e09c2`):

1. **Intentional disable vs accidental missing config** — agreed it matters; the config model today has no explicit disable flag for the Portal OAuth gate, so distinguishing the two would need a new config key. That's a behavior addition beyond this one-line observability fix, so we're leaving it out of this PR.
2. **Precedence observability without credentials** — fair; note the adjacent malformed-client_id branch already includes the client_id in its message (that's pre-existing). Redacting it is a separate change.
3. **Actionable, stable warning + reason code / anti-flood** — reasonable follow-up; `register()` currently runs once at gateway startup, so the flood risk is on reloads. Doesn't block this fix.
4. **Skip-state reset on config transitions** — `LAST_SKIP_REASON` is already reset at the top of every `register()` call; a dynamic unregister-on-remove path would need new plugin-API surface (`unregister_dashboard_auth_provider` doesn't exist today), so we're not adding it here.
5. **Fail-closed** — preserved: no client_id → `register()` returns without registering a provider; no OAuth request is ever made with an empty client id.
6. **Logger scoping consistency** — `nous` uses the same `logging.getLogger(__name__)` pattern as `basic`/`self_hosted`/`drain`; the test change aligns the assertion with the logger that test is actually about.

PR #90781, opened 2026-08-20
fix(telegram): protect **> blockquote marker before bold conversion

sg-shag:
## What does this PR do?

Fixes #90773.

`format_message()` converts blockquotes with a dedicated regex step, but that step ran **after** the bold-conversion step. The bold regex (`\*\*(.+?)\*\*`, non-greedy) matches starting at the very first `**` it finds — so on a line like:

```
**> **Action Plan** 1. **Health Shag Area Skill** - do the thing||
```

it pairs the opening `**` of the `**>` expandable-blockquote marker with the very next `**` (right before "Action"), replacing `**> **` with a placeholder holding `> ` as "bold content" *before* the blockquote step ever runs. By the time the blockquote regex executes, the literal `**>` no longer exists in the text, so the quote is never recognized — Telegram either shows a broken/ungrouped result or corrupted leftover syntax, and the trailing `||` is left dangling with no opener.

### The fix

Move blockquote/expandable-quote marker protection to run **before** bold/italic/strikethrough/spoiler conversion — the same placeholder technique already used for fenced code, inline code, and links. Only the `**>`/`>`/`||` markers themselves are protected; the quoted content is left unescaped so bold/italic/spoiler conversions still apply normally to markdown nested inside the quote.

Also extended `_strip_mdv2()` (the plain-text fallback used when a MarkdownV2 payload is rejected) to strip blockquote/expandable-quote markers, so a fallback never leaks raw `**>`/`>`/`||` syntax to the user — that fallback already handled bold/italic/strikethrough/spoiler but had no blockquote case at all.

### A second, independent defect found during live verification

Testing this fix against the real Bot API with a realistic multi-area digest message surfaced a second bug in the **standalone** `_send_telegram()` path (`tools/send_message_tool.py`, used by `hermes cron run` and any non-gateway-driven delivery): when the formatted message needs more than one chunk, `truncate_message()` appends a raw `" (1/2)"` style suffix *after* MarkdownV2 escaping, with unescaped parentheses. That's invalid MarkdownV2, so Telegram rejects `parse_mode` for **every** chunk and the whole message falls back to plain text — even when the actual entity (e.g. an expandable blockquote) was never split across the chunk boundary and would otherwise have rendered fine. The gateway's own live-agent `send()` already escapes this suffix (`_separate_chunk_indicator_from_fence` + a regex substitution); the standalone cron path never got the same fix. Applied the identical escaping there.

[...]

Issue #90773, opened 2026-08-20
[Bug]: Telegram MarkdownV2 bold conversion runs before blockquote conversion, corrupting `**>` expandable blockquote marker


bongiozzo:
## Describe the bug

`format_message()` in `plugins/platforms/telegram/adapter.py` converts a Telegram MarkdownV2 expandable blockquote (`**> ... ||`) incorrectly whenever the quoted text itself contains `**bold**` markup immediately after the opener. The bold-conversion step runs before the blockquote-conversion step, so the bold regex consumes the leading `**` of the `**>` marker before the blockquote step ever sees it. The result: no expandable (or even regular) blockquote entity reaches Telegram — the raw `**>` / `||` markers, or garbled remnants of them, leak into the message as literal text.

## Root cause

In `format_message()`, the numbered steps run in this order:

- step 5: bold — `re.sub(r'\*\*(.+?)\*\*', ...)`
- step 9: blockquote — `re.sub(r'^((?:\*\*)?>{1,3}) (.+)$', _convert_blockquote, ..., flags=re.MULTILINE)`

Given input:

```
**> **Action Plan for Health** 1. **Health Shag Area Skill** - do the thing||
```

the non-greedy bold regex at step 5 matches starting at the very first `**` (position 0) and closes at the *next* `**` it finds — the one right before "Action" — capturing `"> "` as the "bold" content. That destroys the `**>` prefix (it's replaced by a placeholder token) before step 9's regex ever runs, so the blockquote conversion never fires on that line. The trailing `||` is left dangling with no matching opener.

### Reproduction (pure regex, no bot token needed)

```python
import re

placeholders = {}
counter = [0]
def _ph(v):
    key = f"\x00PH{counter[0]}\x00"
    counter[0] += 1
    placeholders[key] = v
    return key

text = "**> **Action Plan for Health** 1. **Health Shag Area Skill** - do the thing||"

# step 5: bold (as in adapter.py)
text5 = re.sub(r'\*\*(.+?)\*\*', lambda m: _ph(f'*{m.group(1)}*'), text)
print(repr(text5))
# '\x00PH0\x00Action Plan for Health\x00PH1\x00Health Shag Area Skill** - do the thing||'

print("**>" in text5)  # False — the expandable-quote opener is gone
```

[...]

Issue #9656, opened 2026-04-14
Feature request: optionally echo inbound voice-message transcript back to the user


ddagent:
## Feature Description
Add an option for Hermes to echo back the inbound voice-message transcript it actually used, so the user can verify what was heard before relying on the response.

## Motivation
Voice-message STT is sometimes imperfect. A user may speak clearly, but the transcription can still introduce typos, substitutions, or dropped words. Right now Hermes can act on the transcript without the user seeing the exact text that was used.

## Proposed Solution
Add a configurable mode that sends the recognized transcript back to the user when a voice message is received.

Example Telegram UX:
- User sends a voice note
- Hermes replies first with something like:
  "Transcript heard: <transcribed text>"
- Then Hermes continues with the normal answer

This could be:
- always enabled for voice messages, or
- configurable per platform, or
- configurable per chat/user

## Example Use Case
On Telegram, the user sends a voice message and Hermes echoes the transcript back. This lets the user confirm that the speech-to-text model did not introduce typos or transcription errors before trusting the answer.

alt-glitch:
Related: PR #13727 implements voice transcript echo with stt.send_transcription config key.

julibeg:
+1 this would be very useful. I’m running the official Hermes Docker image and can’t patch the source locally. For Discord voice messages, I need a config-only way to echo the STT transcript so that I can verify what Hermes heard before trusting the response.

djangonavarro220:
This is now hitting real users after the upstream voice interrupt fix in #6600 / commit `d55304c39`.

Current behavior: every successful inbound voice-message STT result is echoed visibly to the chat as `🎙️ "<transcript>"` before the normal assistant reply. That is useful for STT debugging, but it is noisy for users who send Telegram voice notes as a convenience and expect Hermes to just answer.

There does not appear to be a config knob for this today. The only available workaround is `stt.enabled: false`, which is too blunt because it disables automatic voice-note understanding entirely.

Suggested config shape, bikeshed names welcome:

```yaml
stt:
  enabled: true
  echo_transcript: false   # default should probably be false, or at least configurable per platform/chat
```

Or platform-scoped if you prefer avoiding a global behavior change:

```yaml
display:
  platforms:
    telegram:
      echo_voice_transcripts: false
```

Important distinction: transcribing and injecting the transcript into the agent-visible message should remain enabled. Only the visible receipt/echo should be optional.

teknium1:
[...]

PR #7369, opened 2026-04-10
feat(gateway): stream reasoning as collapsible blockquote on Telegram

n-WN:
## Summary

When `display.show_reasoning` is enabled and the model returns thinking blocks, the gateway now streams thinking content in real-time and renders it as a Telegram expandable blockquote — all in a single message.

Previously, the gateway never registered `reasoning_callback`, so thinking deltas were silently dropped during streaming. The `show_reasoning` post-processing fallback was also skipped when `already_sent=True` (streaming mode), leaving users with zero thinking visibility and a perceived long TTFT.

## User-visible behavior

```
[t=0s]  💭 Thinking…
        The user asked about ▉           ← thinking streams in real-time

[t=2s]  💭 Thinking                      ← collapses into expandable blockquote
        ▶ (tap to expand full thinking)
        
        The answer is ▉                  ← response text starts streaming

[t=3s]  💭 Thinking                      ← stays collapsed
        
        The answer is **42**. This       ← response continues streaming
        is because...
```

Single message throughout. Both thinking and response are streamed progressively.

### `gateway/stream_consumer.py`

- Add `_html_prefix` attribute and `_in_thinking_phase` flag
- Add `start_thinking()` — marks the consumer as in thinking phase
- Add `switch_to_html_mode(prefix)` — atomically sets the HTML expandable blockquote prefix and clears the accumulated thinking text from the buffer
- Add `_send_or_edit_html(text)` — bypasses the adapter's MarkdownV2 path and talks directly to the Telegram bot API in HTML mode, prepending `_html_prefix` to every send/edit. Handles flood control and edit failures gracefully.
- `_send_or_edit()` entry point routes to `_send_or_edit_html()` when `_html_prefix` is set

### `gateway/run.py`

- Register `reasoning_callback` on the agent when `show_reasoning` is enabled and a stream consumer is active
- `_reasoning_cb`: feeds thinking deltas to the stream consumer via `on_delta()` for real-time display, prefixed with `💭 Thinking…` on first delta
- `_wrapped_stream_delta_cb`: on first text delta, formats accumulated thinking as `<blockquote expandable>` HTML and calls `consumer.switch_to_html_mode()`, then forwards text deltas normally
- Pass `_reasoning_already_sent` flag through `agent_result` so the post-processing fallback path skips re-sending when streaming already delivered reasoning
- Non-streaming fallback: sends thinking + response as a single HTML message with expandable blockquote via the Telegram bot API
- Non-Telegram platforms: falls back to plain text prepend

[...]

PR #98419, opened 2026-08-30
feat(gateway): collapse Telegram STT echoes into expandable HTML quotes

Artemonim:
## What does this PR do?

The gateway already echoes inbound voice STT when `stt.echo_transcripts` is true. On Telegram that echo is a full `🎙️ "…"` line, so a long note dumps a wall of text into the chat, and markdown characters in the speech can break MarkdownV2 parse.

This still echoes. On Telegram the body is an HTML expandable blockquote (`<blockquote expandable>`), so the transcript sits in a collapsed quote. Other platforms keep the classic line.

I used HTML instead of the MarkdownV2 `**> … ||` expandable marker because that marker is currently eaten by bold conversion ([#90773](https://github.com/NousResearch/hermes-agent/issues/90773), open PR [#90781](https://github.com/NousResearch/hermes-agent/pull/90781)). Speech is HTML-escaped and sent with `metadata["telegram_html"]`, so it never goes through `format_message()`.

`stt.echo_transcripts` stays `true`. Set it to `false` if you don't want the echo at all.

Transcripts longer than Telegram's cap are re-wrapped so every continuation is still a complete collapsed quote, with `(N/M)` outside the tags.

## Related Issue

No dedicated issue for collapsed-quote wrapping. Adjacent, not this:

- [#9656](https://github.com/NousResearch/hermes-agent/issues/9656) (closed): the echo itself, already on main
- [#90773](https://github.com/NousResearch/hermes-agent/issues/90773) / [#90781](https://github.com/NousResearch/hermes-agent/pull/90781): MarkdownV2 `**>` is why this path is HTML
- [#7368](https://github.com/NousResearch/hermes-agent/issues/7368) / [#7369](https://github.com/NousResearch/hermes-agent/pull/7369): collapsible quote for reasoning, not transcripts

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `gateway/stt_echo.py`: format Telegram echoes as expandable HTML; keep `🎙️ "…"` elsewhere; re-wrap long quotes with `(N/M)` outside the tags
- `gateway/run.py`: both echo send sites (immediate + pending) use `format_stt_transcript_echo` / `stt_echo_metadata`
- `plugins/platforms/telegram/adapter.py`: `telegram_html` → `_send_html_message` (skip MarkdownV2); STT-aware HTML chunking; DM-topic fail-closed routing so a dead topic cannot strip `message_thread_id` and leak the echo into the general chat
- `hermes_cli/config_defaults.py`, `website/docs/user-guide/configuration.md`: comment only (Telegram collapsed quote vs classic line)
- Tests: `tests/gateway/test_stt_echo_format.py`, `test_telegram_format.py` (split keeps the quote on every chunk), `test_telegram_voice_v0_regressions.py`, `test_telegram_thread_fallback.py` (DM-topic fail-closed / reply-anchor / thread-not-found)

## How to Test

1. `scripts/run_tests.sh tests/gateway/test_stt_echo_format.py tests/gateway/test_telegram_format.py tests/gateway/test_telegram_voice_v0_regressions.py tests/gateway/test_telegram_thread_fallback.py tests/gateway/test_stt_transcript_echo_config.py`
2. Telegram: send a short voice note with `stt.echo_transcripts: true`. The echo should be a collapsed quote, not a full `🎙️ "…"` line.
3. Speak something with `*` / `_` / backticks. The send should not fall back to plain text from a MarkdownV2 parse error.
4. Long transcript (or temporarily lower `MAX_MESSAGE_LENGTH` in a test): each continuation still has `<blockquote expandable>`, with `(N/M)` after `</blockquote>`.
5. Optional: voice in a Telegram DM topic. Echo should stay in the topic (reply-anchor). Missing anchor or a dead topic should fail closed, not land in the general chat.

I have not live-tested against the Bot API in this PR; coverage is the pytest files above.

### Code

- [x] I've run `scripts/run_tests.sh` on the files above (76 passed). Template's `pytest tests/ -q` is the full suite; I did not run the entire tree.
- [x] I've tested on my platform: Windows 11 (unit tests; live Telegram)

- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — or N/A (comment only in `config_defaults.py`; no new key)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (formatting / Telegram API, no OS process paths)


liyangbing:
<p>This is a thoughtful formatting fix: Telegram STT echoes stay available but become a collapsed HTML quote, while non-Telegram platforms retain their existing plain-text form. Re-wrapping every long chunk and preserving the DM-topic fail-closed routing are particularly important details.</p>

<h3>Suggested contract</h3>

<ul>
<li>Preserve transcript data exactly on the HTML-fallback path. The current fallback unescapes the chunk and then removes tag-shaped text with a regular expression, so spoken literal content such as <code>&lt;tag&gt;</code> can be deleted. Strip only the wrapper markup, then unescape the remaining transcript without treating it as HTML.</li>
<li>Make the Bot API/client compatibility contract explicit. Expandable blockquotes require a supported Telegram Bot API/client combination; if HTML is rejected, report that collapse was lost while preserving the full escaped transcript and its topic routing.</li>
<li>Keep every chunk within the UTF-16 message limit, including the wrapper, transcript content, and <code>(N/M)</code> indicator; test very small configured limits and long unbroken words, not only the normal 4096 boundary.</li>
<li>Ensure chunking preserves transcript order and content, including leading/trailing whitespace, newlines, Unicode surrogate-pair accounting, and HTML entities; a split should never silently trim meaningful speech.</li>
<li>Keep the HTML path equivalent to the normal send path for reply anchors, topic IDs, thread-not-found handling, notification/link-preview options, and partial-send failure reporting.</li>
<li>Make the preformatted-HTML metadata capability-specific and fail closed if an unrelated caller supplies malformed or untrusted HTML; escaped transcript content must never become executable markup.</li>
</ul>

<p>The focused tests cover entity-safe splitting and DM-topic routing, but the fallback branch deserves a regression test with a transcript containing literal angle brackets, ampersands, quotes, and text that resembles a blockquote tag. Since the PR states that the Bot API was not live-tested, a small real Telegram sandbox or recorded API-contract witness would also validate parse-mode behavior and the expandable attribute.</p>

<h3>Regression matrix</h3>

<ul>
<li>short, multiline, Unicode, Markdown-like, entity-like, and literal HTML-looking transcripts</li>
<li>HTML accepted, HTML parse rejected, network timeout, rate limit, and partial multi-chunk failure</li>
<li>normal limit, UTF-16 boundary, very small limit, long unbroken token, and entity boundary</li>
<li>short one-chunk quote and long multi-chunk quotes with indicator outside every wrapper</li>
<li>public chat, ordinary thread, private DM topic with anchor, missing anchor, and dead topic</li>
<li>Telegram clients/API versions that support or do not support expandable blockquotes</li>
<li>asserting exact transcript preservation and no fourth-party HTML/tag interpretation in plain fallback text</li>
</ul>

<p>Evidence reviewed: <code>gateway/stt_echo.py</code>, <code>plugins/platforms/telegram/adapter.py</code>, <code>gateway/run.py</code>, and the focused Telegram/STT tests in PR head <code>de66b27046fa8b446bd557b0571221e155a2f0b2</code>.</p>

Artemonim:
@liyangbing The fallback path does not drop spoken `<tag>`. Transcripts are `html.escape`d before wrapping, so the body is `&lt;tag&gt;` and the tag regex never sees a real `<`. Order is: strip the blockquote wrappers, strip leftover literal tags, then unescape. A regression test would pin that invariant; it is not a data-loss bug on this path.

The rest of the suggested contract is either already true here or a larger follow-up than this formatting change:

- Expandable quotes need Bot API 7.3+. That is in the PR body. If HTML parse fails we log and send plain text. We do not also tell the user "collapse was lost"; the echo is optional and already non-fatal.
- Chunks are sized with `utf16_len`, including the wrapper and `(N/M)`. Tests use 400 and 180, not only 4096. Entity-safe splits are covered. A continuation can lose leading whitespace the same way `truncate_message` does.
- DM-topic fail-closed (missing anchor / dead topic) is the third commit. Retry/timeout/partial-chunk parity with the full `send()` loop is out of scope.
- `telegram_html` is only set by the STT echo formatter. A general HTML send gate for untrusted callers is a different change.

_Answered by Cursor Grok 4.6 High_

kvnloo:
Exact-head KEEP on NousResearch/hermes-agent#98419 head 7ab31c7466.

KEEP collapsing Telegram STT echoes into an HTML expandable blockquote when `stt.echo_transcripts` is true, instead of a full 🎙️ markdown wall that can break MarkdownV2. KEEP other platforms on the classic echo line. KEEP tests under `tests/gateway/test_stt_echo_format.py` / telegram format + voice regressions + docs.

CHECK: HTML path only on Telegram. CHECK: long transcripts stay collapsed by default. Author Artemonim not kvnloo. No competing PR.

PR #84510, opened 2026-08-12
fix(auxiliary): walk configured fallback chains

everninemedia:
## Summary
- walk every configured auxiliary fallback candidate instead of stopping at the first recoverable provider error
- add per-task fallback_to_main=false to prevent hidden main-model spend
- cover sync and async multihop chains plus the no-main contract

## Verification
- uv run --extra dev pytest tests/agent/test_auxiliary_client.py -q
- 184 passed

This preserves backward compatibility: fallback_to_main defaults to true.

PR #98422, opened 2026-08-30
fix(auxiliary): continue through configured fallback chain

herbalizer404:
## Problem

Auxiliary fallback chains can stop before later configured entries in two distinct failure modes.

### Explicit errors

If the primary auxiliary provider fails with a recoverable error and `fallback_chain[0]` also fails (for example both return HTTP 402), later entries are not attempted.

```text
primary ollama-cloud/minimax-m3 -> 402
fallback_chain[0] ollama-free/minimax-m3 -> 402
fallback_chain[1] openai-codex/gpt-5.6-luna -> never attempted
```

### Progressing stalls

Compression has a host-level progress-aware timeout because a connection can stay open without completing. The stall path previously pinned only the first usable fallback entry. If both the primary and `fallback_chain[0]` streamed until their total ceilings without producing a committed compression, Hermes degraded without attempting `fallback_chain[1]`.

Production evidence showed:

```text
ollama-cloud/nemotron-3-super -> 600s ceiling
ollama-free/nemotron-3-super  -> 600s ceiling
openai-codex/gpt-5.6-luna     -> never attempted
compression tokens            -> unchanged
```

## Root cause

- `_try_configured_fallback_chain()` returned the first constructible candidate. Sync/async fallback calls then re-raised recoverable non-auth failures, so execution never resumed at the next configured index.
- `resolve_compression_fallback_route()` returned only the first structurally usable stall fallback, and `_retry_compression_on_fallback_chain()` performed only one pinned retry before degrading.

### Explicit-error path

- Add an explicit `start_index` when resolving configured fallback candidates.
- Continue through later configured entries after recoverable fallback-worthy failures.
- Preserve failure-scope semantics, credential refresh, per-entry timeout, cache replanning, unhealthy-provider handling, route telemetry, sync/async parity, and existing safety nets.
- Honor HTTP 402 cooldown state while walking configured fallbacks.

### Stall path

- Resolve every structurally usable compression fallback route in declared order.
- Run one bounded, pinned retry per entry, with a fresh commit fence and that entry's timeout semantics.
- Stop at the first committed compression; degrade exactly once only after all entries are exhausted.
- Re-check hard cancellation between retries and disable nested stall fallback recursion.
- Scope and restore both summary-route `ContextVar` values so consumed-route state cannot leak across attempts or tests.

## Tests

Regression coverage includes:

- sync and async primary 402 -> fallback 0 402 -> fallback 1 success;
- complete explicit-error chain exhaustion preserving the primary error;
- stale fallback authentication continuing to later entries;
- primary stall -> malformed entries skipped -> fallback 0 no-op -> fallback 1 success;
- ordered duplicate entries, fresh fences, entry-specific timeout/ceiling budgets, and one final degrade;
- hard cancellation before and between retries;
- real `ContextCompressor -> call_llm` kwargs for the second pinned route;
- `ContextVar` isolation with the ownership suite in the same pytest process.

Verified on the rebased PR head:

```text
207 passed in tests/agent/test_*compression*.py
189 passed in tests/agent/test_auxiliary_client.py
python -m py_compile (all touched Python files)
git diff --check
```

Independent review: native GPT-5.6 Luna (high) — APPROVE. Second reader disabled by user.

## Impact

Users can rely on every entry in `auxiliary.<task>.fallback_chain` being attempted in order for recoverable API failures and for compression routes that hit their progress-aware total ceiling. Later fallbacks such as Luna are no longer silently skipped after two stalled Ollama routes.

### DIFF
--- a/agent/auxiliary_client.py
+++ b/agent/auxiliary_client.py
@@ -5275,6 +5275,29 @@ def _fallback_provider_from_label(label: str) -> str:
     return match.group(1).strip() if match else str(label or "").strip()
 
 
+def _next_configured_fallback_index(label: str) -> int:
+    """Return the configured-chain index immediately after ``label``."""
+    match = re.match(r"fallback_chain\[(\d+)\]", label or "")
+    return int(match.group(1)) + 1 if match else 0
+
+
+def _fallback_candidate_failure_reason(exc: Exception) -> Optional[str]:
+    """Classify errors that may advance to another auxiliary fallback."""
+    if _is_auth_error(exc):
+        return "auth error"
+    if _is_payment_error(exc):
+        return "payment error"
+    if _is_rate_limit_error(exc):
+        return "rate limit"
+    if _is_model_incompatible_error(exc):
+        return "model incompatible with route"
+    if _is_invalid_aux_response_error(exc):
+        return "invalid provider response"
+    if _is_connection_error(exc):
+        return "connection error"
+    return None
+
+
 class _FallbackDestination(NamedTuple):
     provider: str
     base_url: str
@@ -5876,6 +5899,7 @@ def _try_configured_fallback_chain(
     failed_provider: str,
     reason: str = "error",
     failed_model: Optional[str] = None,
+    start_index: int = 0,
 ) -> Tuple[Optional[Any], Optional[str], str]:
     """Try user-configured fallback_chain for a specific auxiliary task.
 
@@ -5934,7 +5958,8 @@ def _try_configured_fallback_chain(
     tried = []
     min_ctx = _task_minimum_context_length(task)
 
-    for i, entry in enumerate(chain):
+    start_index = max(0, start_index)
+    for i, entry in enumerate(chain[start_index:], start=start_index):
         if not isinstance(entry, dict):
             continue
         fb_provider = str(entry.get("provider", "")).strip()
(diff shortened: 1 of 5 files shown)


alt-glitch:

Related to #84510: both make configured auxiliary fallback chains advance after recoverable candidate failures, using different loop and exhaustion-policy implementations. Please consolidate before merge.

liyangbing:
<p>This addresses a real reliability failure: a configured auxiliary chain should not silently stop after the first constructible client. The explicit continuation index, sync/async regression pair, and stale-credential case make the intended behavior concrete.</p>

<h3>Suggested contract</h3>

<ul>
<li>Advance by the configured entry identity, not only by a display-label parser; duplicate providers, reordered labels, and a future label-format change must not retry the same candidate or skip a later one.</li>
<li>Guarantee progress or termination on every loop iteration. If resolution returns no candidate, an invalid candidate, or a label that cannot map to the chain, exit to the documented safety net rather than restarting at index zero.</li>
<li>Keep failure scope exact: auth/payment failures may quarantine a provider, while model/transport failures should skip only the failed model identity when another model on that provider remains viable.</li>
<li>Do not swallow cancellation, operator abort, malformed configuration, or other non-recoverable exceptions merely because a later fallback exists; only the documented recoverable classes should advance the chain.</li>
<li>Preserve the original primary exception and warning after genuine configured-chain exhaustion, while recording every attempted route and the reason it was skipped.</li>
<li>Keep sync and async behavior equivalent for credential refresh, client-cache eviction, per-entry timeout, unhealthy-provider state, route telemetry, and the final main-agent/discovery safety nets.</li>
</ul>

<p>The current tests prove a two-entry 402 chain, stale fallback authentication, and exhaustion, but the index/label contract deserves a direct invariant test. In particular, assert that a repeated provider with different models reaches the second model, that a resolver returning <code>None</code> advances, and that a non-recoverable exception does not invoke a later candidate.</p>

<h3>Regression matrix</h3>

<ul>
<li>sync and async primary failure followed by fallback 0 failure and fallback 1 success</li>
<li>duplicate providers with distinct models, duplicate entries, and reordered chains</li>
<li>auth, payment, rate-limit, connection, incompatible-model, invalid-response, timeout, cancellation, and malformed-config failures</li>
<li>stale credentials after refresh failure, cached-client eviction, and provider unhealthy marking</li>
<li>all configured entries exhausted, preserving the original primary error and warning</li>
<li>empty chain, non-dict entries, invalid providers/models, and a resolver that returns no candidate</li>
<li>route telemetry and user-visible diagnostics showing attempted, skipped, and selected candidates consistently</li>
</ul>

<p>Evidence reviewed: <code>agent/auxiliary_client.py</code> and <code>tests/agent/test_auxiliary_client.py</code> in PR head <code>90a0e26c5d2fc54b97f2c96cc3446b2b393bb107</code>.</p>

herbalizer404:
Follow-up pushed on top of the refreshed `main` branch.

The production change is intentionally narrow: configured fallback entries and an explicit auxiliary primary now honor the existing payment-error cooldown. After the first 402, repeated compression chunks skip those already-known-unhealthy routes and go directly to the next configured fallback. No model selection or configuration policy changed.

Regression coverage now exercises two consecutive calls in both sync and async modes: primary 402 -> fallback[0] 402 -> fallback[1] success, followed by a second call that reaches fallback[1] without retrying the first two routes.

Validation: `scripts/run_tests.sh tests/agent/test_auxiliary_client.py -q` -> 189 passed; Ruff and `git diff --check` pass.

herbalizer404:
Closing at the author request; the local compression fallback patches have been reverted.

PR #38816, opened 2026-06-04
fix(skills): treat user-configured taps as 'tap' trust level, not 'community'

liuhao1024:
## What does this PR do?

Skills installed from user-configured taps (`hermes skills tap add owner/repo`) are now treated with a new "tap" trust level instead of "community". This means:
- **Safe/caution verdicts**: allowed without `--force` (same as "trusted")
- **Dangerous verdicts**: blocked by default, but `--force` CAN override (unlike "community" and "trusted" where dangerous is always a hard block)

Previously, tap-sourced skills were indistinguishable from random community skills, making it impossible to install operational skills (DB queries, API integrations, deployment scripts) from trusted private repos without manual `rsync` workarounds.

## Related Issue

Fixes #38801

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `tools/skills_guard.py`: Added `"tap"` entry to `INSTALL_POLICY` with `("allow", "allow", "block")` semantics; updated `should_allow_install()` comments to document that `--force` overrides dangerous for "tap" but not "community"/"trusted"
- `tools/skills_hub.py`: Modified `GitHubSource.__init__` to build `_tap_repos` set from `extra_taps`; modified `trust_level_for()` to return `"tap"` for user-configured tap repos; added `"tap"` to dedup trust rank
- `hermes_cli/skills_hub.py`: Added `"tap": "cyan"` to all trust_style display mappings and `"tap": 2` to `_TRUST_RANK` dicts
- `tests/tools/test_skills_guard.py`: Added `TestTapTrustLevel` class with 7 tests covering all trust/verdict/force combinations
- `tests/tools/test_skills_hub.py`: Added 3 tests to `TestTrustLevelFor` verifying tap repo trust resolution

## How to Test

1. Configure a tap: `hermes skills tap add paicha/skills`
2. Install a skill from it: `hermes skills install growth-dashboard` — should succeed for safe/caution verdicts
3. For a skill with dangerous verdict: `hermes skills install <skill>` → blocked, then `hermes skills install <skill> --force` → should succeed
4. Run tests: `pytest tests/tools/test_skills_guard.py tests/tools/test_skills_hub.py -x -q` → 189 passed

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
[...]

PR #98423, opened 2026-08-30
fix(skills): trust operator-added custom taps

GoBeromsu:
## Summary

- classify repositories added through `hermes skills tap add` as an `operator` trust tier
- allow safe/caution/dangerous operator-tap bundles while preserving quarantine, scanning, verdicts, findings, and reports
- reapply current source-router trust after cached scans so tap removal immediately restores community policy
- leave built-in/default taps and untapped community sources unchanged

## Rationale

Adding a custom tap is already an explicit operator trust decision. Today those repositories remain `community`, so prose-oriented false positives can permanently block normal `hermes skills update` even though the operator deliberately registered the source. This change reuses the existing `SkillBundle.trust_level` / `ScanResult.trust_level` pipeline rather than adding another authorization database or prompt.

## Safety

- exact case-insensitive owner/repo equality; no prefix matching
- only constructor-supplied `extra_taps` qualify; `DEFAULT_TAPS` do not
- no publisher/owner hardcode
- scan verdicts and findings are not suppressed or downgraded
- current tap membership is reapplied on each install/update, including cached scans

## Tests

```text
venv/bin/python -m pytest tests/tools/test_skills_guard.py tests/tools/test_skills_hub.py tests/hermes_cli/test_skills_hub.py -q
153 passed
```

### DIFF
--- a/hermes_cli/skills_hub.py
+++ b/hermes_cli/skills_hub.py
@@ -717,6 +717,12 @@ def do_install(identifier: str, category: str = "", force: bool = False,
         source_url=source_url_for_bundle(bundle),
         cache_dir=HUB_DIR / "scan-cache",
     )
+    # Source identifiers drive scanner provenance, while the active source
+    # router owns trust classification. Reapply current bundle trust on every
+    # install/update so cached scans cannot retain stale tap membership.
+    result.trust_level = bundle.trust_level
+    result.scan_provenance["trust_level"] = bundle.trust_level
+    scan_provenance["trust_level"] = bundle.trust_level
     c.print(format_scan_report(result))
     freshness = "fresh" if scan_provenance["fresh"] else "cached"
     c.print(
--- a/hermes_cli/skills_hub.py
+++ b/hermes_cli/skills_hub.py
@@ -398,7 +398,9 @@ def do_browse(page: int = 1, page_size: int = 20, source: str = "all",
 
     # Collect results from all (or filtered) sources in parallel.
     # Per-source limits are generous — parallelism + 30s timeout cap prevents hangs.
-    _TRUST_RANK = {"builtin": 3, "trusted": 2, "community": 1}
+    _TRUST_RANK = {
+        "builtin": 4, "operator": 3, "trusted": 2, "community": 1,
+    }
     # NOTE: when the centralized index is available, parallel_search_sources
     # skips the external API sources and serves everything from "hermes-index".
     # That source MUST therefore carry a limit large enough to cover the whole
@@ -925,7 +927,9 @@ def browse_skills(page: int = 1, page_size: int = 20, source: str = "all") -> di
     )
 
     page_size = max(1, min(page_size, 100))
-    _TRUST_RANK = {"builtin": 3, "trusted": 2, "community": 1}
+    _TRUST_RANK = {
+        "builtin": 4, "operator": 3, "trusted": 2, "community": 1,
+    }
     # "hermes-index" must carry a high limit: when the index is available the
     # router skips external API sources and serves everything from it, so a
     # low cap here silently truncates the whole hub (see do_browse note).
--- a/hermes_cli/skills_hub.py
+++ b/hermes_cli/skills_hub.py
@@ -28,6 +28,15 @@ from hermes_constants import display_hermes_home
 _console = Console()
 
 
+def _apply_current_bundle_trust(
+    result, scan_provenance: dict, trust_level: str
+) -> None:
+    """Bind policy to the active source router, never stale scan-cache trust."""
+    result.trust_level = trust_level
+    result.scan_provenance["trust_level"] = trust_level
+    scan_provenance["trust_level"] = trust_level
+
+
 def _display_source(r) -> str:
     """Human-facing source label for a result row.
 
@@ -720,9 +729,7 @@ def do_install(identifier: str, category: str = "", force: bool = False,
     # Source identifiers drive scanner provenance, while the active source
     # router owns trust classification. Reapply current bundle trust on every
     # install/update so cached scans cannot retain stale tap membership.
-    result.trust_level = bundle.trust_level
-    result.scan_provenance["trust_level"] = bundle.trust_level
-    scan_provenance["trust_level"] = bundle.trust_level
+    _apply_current_bundle_trust(result, scan_provenance, bundle.trust_level)
     c.print(format_scan_report(result))
     freshness = "fresh" if scan_provenance["fresh"] else "cached"
     c.print(
--- a/tests/hermes_cli/test_skills_hub.py
+++ b/tests/hermes_cli/test_skills_hub.py
@@ -48,6 +48,85 @@ def test_current_bundle_trust_overrides_fresh_or_cached_scan(
     assert result.verdict == "dangerous"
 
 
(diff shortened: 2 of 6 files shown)


alt-glitch:

Related to #38816: both repair custom-tap trust classification, but this approach automatically permits dangerous scan verdicts for operator-added taps while #38816 retains an explicit force gate. Please consolidate the policy before merge.

liyangbing:
<p>This is a useful policy clarification: an operator who explicitly adds a custom tap should not have ordinary prose-oriented scanner findings permanently block that source, while the scan verdicts and findings themselves must remain visible. Reapplying the current trust level when consuming a cached scan is the right direction for tap removal and reconfiguration.</p>

<h3>Suggested contract</h3>

<ul>
<li>Keep trust classification separate from scanner output: <code>operator</code> may change the install decision, but must not rewrite, suppress, or downgrade the recorded verdict, findings, provenance, or report.</li>
<li>Define the scope precisely. The current set is repository-wide by exact owner/repo; document whether a tap's configured path is intentionally not part of the trust boundary, and test a second path in the same repository.</li>
<li>Make normalization explicit for case, leading/trailing slashes, duplicate separators, malformed entries, and non-string tap values; malformed configuration should be ignored or rejected deterministically rather than raising during source construction.</li>
<li>On tap removal, source replacement, or path reconfiguration, recompute trust from current membership before both install and update, including cached scans and already-discovered bundles.</li>
<li>Keep the operator policy bounded to this explicit install decision. It must not silently grant publisher trust, bypass authorization prompts, alter runtime permissions, or make a community source trusted elsewhere.</li>
<li>Make dangerous operator installs auditable: show the source, normalized repository, verdict, findings, and the explicit operator-trust reason in the report and event log.</li>
</ul>

<p>Because <code>operator</code> allows a dangerous verdict, the most important negative test is that the same repository is blocked again immediately after its tap is removed, without a stale cached trust level surviving. Please also cover a cached scan whose provenance says <code>community</code> but whose current source router says <code>operator</code>, and the reverse transition.</p>

<h3>Regression matrix</h3>

<ul>
<li>safe, caution, and dangerous findings from an active custom tap</li>
<li>exact owner/repo match versus a sibling repository, altered case, extra path, and malformed identifier</li>
<li>custom tap path A versus another skill path in the same repository</li>
<li>default/built-in tap and ordinary community source with identical findings</li>
<li>fresh scan, cached scan, cache hit after tap addition, and cache hit after tap removal</li>
<li>duplicate taps, empty or malformed tap records, and repository normalization variants</li>
<li>install and update reports preserving verdicts/findings while exposing the operator decision</li>
</ul>

<p>Evidence reviewed: <code>tools/skills_guard.py</code>, <code>tools/skills_hub.py</code>, <code>hermes_cli/skills_hub.py</code>, and their focused tests in PR head <code>92202db651a19fac6982b40eaf3b81929fd400fb</code>.</p>

andrexibiza (commented):
I like the direction of separating **current source trust** from scanner findings, especially the attempt to recompute trust after a cached scan so removing a tap cannot leave a stale elevated decision behind. The exact owner/repo normalization is also materially safer than prefix matching.

I reviewed exact head `92202db651a19fac6982b40eaf3b81929fd400fb` against its direct parent/current `main@26350357d76e4508c8df9304a3374bdc5a6f6220`, the five-file diff, the current install/update path, the scanner cache/provenance contract, the existing discussion here, and overlapping #38816. There is one security boundary I do not think can land in the current shape, plus one provenance/merge-topology issue that should be fixed in the same convergence.

### P1 — adding a tap cannot be a perpetual approval for future `dangerous` revisions

`tools/skills_guard.py` changes the new `operator` row to:

```py
"operator": ("allow", "allow", "allow")
```

and the regression explicitly requires `dangerous` to pass without `--force`.

That makes a source-registration decision authorize a different, later mutation: executing whatever code that repository publishes in the future, even when Hermes' own current scanner classifies the exact fetched bundle as `dangerous`. The lifetime problem is especially concrete on update: current `hermes_cli/skills_hub.py::do_update()` calls `do_install(..., force=True, ...)` unconditionally as an internal overwrite mechanism. So a benign bundle can be tapped and installed today, the upstream can later be compromised/change to a scanner-`dangerous` revision, and normal `hermes skills update` has no fresh dangerous-content consent boundary at all.

That is the other side of this trust shape: **trusting a repository for discovery/update is not proof that every future realization of that repository is approved to cross the dangerous-code gate.** A tap is durable source authority; a scan verdict is exact-bundle evidence. They need separate lifetimes.

Required contract: keep the useful `operator`/tap behavior for safe+caution, but make a current `dangerous` realization require an explicit *current* operator decision. The existing `do_update(...)->do_install(force=True)` flag cannot count as that decision because it is set internally on every update. Split the concepts (for example, overwrite/update permission vs dangerous-policy override) and add the adversarial regression:

1. add operator tap;
2. install a safe bundle;
3. change the fetched bundle to a `dangerous` verdict;
4. ordinary update must block before install;
5. only an explicit user-provided dangerous override may proceed;
6. removing the tap restores community policy immediately.

This also means #38816 is not simply "the stricter alternative": its proposed tap-dangerous `--force` escape hatch still interlocks with the existing unconditional `force=True` in `do_update()`. Whichever PR becomes the convergence object needs to repair that call-site semantics too.

### P2 / proof-boundary — do not rewrite the scan receipt to represent current routing trust

`scan_skill_cached()` records a provenance object containing `bundle_hash`, `scanner_version`, `verdict`, `trust_level`, `findings`, and `scanned_at`; on cache hit it returns that stored object with `fresh=False`. This PR then mutates both `result.scan_provenance["trust_level"]` and `scan_provenance["trust_level"]` to the *current* `bundle.trust_level` before reporting/decision.

Recomputing current trust is correct; rewriting historical scan provenance is not. It turns a receipt about what was scanned under what classification into a mixed-time object. Please preserve the cached scan receipt intact and carry current decision trust separately (`current_trust_level` / `decision_trust_level`, or equivalent). `result.trust_level` can be the live policy input if that is the intended contract, but the provenance record should remain truthful about the scan object it describes.

A useful regression is the two-way cache transition already implied by this PR: cached `community` scan + current operator tap, then cached operator-era scan + tap removed. The install decision should follow current routing trust while the immutable scan provenance still reports the original scan receipt in both cases.

### Interlock / attribution

#38816 by `liuhao1024` is direct prior/competing work on the same `INSTALL_POLICY`, `GitHubSource` trust classifier, CLI trust rendering, and tests. #98423 has useful unique work — exact casefolded repo classification and live trust reapplication across cached scans — so I would treat this as a **superseding convergence candidate**, not a duplicate. Please preserve #38816's authorship/design lineage explicitly while converging on one policy owner; the two PRs should not land independently with contradictory `tap`/`operator` dangerous semantics.

### Exact-object verification

The focused `153 passed` receipt in the PR body is useful construction evidence, but hosted landing evidence has not executed for this head: CI `33299411112`, Docker `33299410801`, and Nix `33299410800` all concluded `action_required`; the CI run currently has **0 jobs**. The repaired/converged head needs fresh exact-head hosted receipts before merge.

There is good work here worth keeping: exact repo equality, default-tap exclusion, and the instinct to re-evaluate current trust instead of trusting stale cache state are all the right ingredients. The remaining fix is to bind that trust to the right lifetime and keep the scan receipt immutable.

Issue #97963, opened 2026-08-29
[Bug] Hygiene compression turn-hold (10s default) makes auto-compress fail 100% for thinking models with long sessions — regression from #92318


whengu:
## Summary

After today's update (merged in `#92318` / `543c85acbf`, `aff5125f8e`), **session hygiene auto-compression fails 100% of the time** for long sessions using thinking models. The new `hygiene_max_turn_hold_seconds: 10` budget (default 10s) is killed before the summary stream produces any `content`, because thinking models (e.g. DeepSeek-V4-Flash-0731) emit a long `reasoning_content` prefix before the first content token. With the 10s wall-clock hold, compression can never finish — every auto-compress attempt fails — and context grows unbounded until the session must be abandoned.

## Reproduction

Environment: Windows, gateway + Feishu, main model `dic/DeepSeek-V4-Flash-0731` (a reasoning model) through an OpenAI-compatible gateway (9Router). Long session (~400–500k tokens as already observed in our logs).

1. Let a session grow past the compression threshold (session > threshold, e.g. 400k/512k).
2. Send any user message so pre-turn hygiene compression runs.
3. Observe (gateway.log):
   ```
   WARNING gateway.run: Session hygiene auto-compress failed: turn-hold budget 10.0s elapsed after 10.0s
   ```
   followed by repeated compressor failures:
   ```
   WARNING agent.context_compressor: Failed to generate context summary: Context compression LLM returned empty content (provider=custom model=dic/DeepSeek-V4-Flash-0731). Further summary attempts paused for 30 seconds.
   ```

## Why the 10s budget is structurally insufficient for thinking models

Measured on the live endpoint (streaming, `stream: true`, same model, **only ~10k prompt tokens** — the real compression prompt is tens of times larger):

```
total 5.7s | first content token at 4.98s | 601 SSE chunks (most are reasoning/empty deltas)
```

A real compression request carries the whole conversation (hundreds of thousands of tokens). Prefill + reasoning before the first content token is inherently far beyond 10 seconds. The turn-hold budget therefore fires **before the model has even started writing the summary**, every single time. The old mechanism (`hygiene_total_ceiling_seconds` 600s / `compression.context_timeout_seconds` 300s) was long enough to let a streaming summary finish; the new fixed 10s hold **silently truncates the entire compression feature for thinking-model deployments** while claiming to "bound" a freeze that the streaming path had already resolved.

[...]

Issue #96953, opened 2026-08-28
[Bug]: Compression commit-fence cancellation + session-hygiene retry livelock — gateway unresponsive ~30 min after /restart on large sessions


lih54767-coder:
## Bug

Session hygiene auto-compression and the compression commit fence interact as a **livelock** on large sessions after `/stop` or `/restart`: every compression attempt is cancelled at the commit fence by the restart/stop invalidation, hygiene immediately re-attempts and waits (up to a 600 s ceiling per round), and inbound messages are demoted to `queue` behind compression — so the gateway goes unresponsive for **~31 minutes total** while zero compression work gets done.

## Environment

- Hermes Agent v0.20.6 (2026.8.27), upstream commit 10b38830, install method: git
- Windows 11, Python 3.11.15, OpenAI SDK 2.24.0
- Feishu gateway, single long-lived DM session

## Steps to reproduce

1. Let a gateway session grow large (~520 messages / ~176K tokens).
2. Issue `/stop` and/or `/restart` (cancellation signals).
3. Session hygiene detects the oversized session on/after restart and triggers auto-compression.
4. Send messages from the messaging platform while compression is in flight.

## What happened (real logs, 2026-08-28)

Hygiene triggered auto-compression at 14:18:38 and retried twice more (14:20:01, 14:28:50), each attempt running up against the 600 s wait ceiling:

```text
14:18:38 INFO gateway.run: Session hygiene: 521 messages, ~175,873 tokens (actual) — auto-compressing (threshold: 85% of 1,000,000 = 850,000 tokens)
14:19:10 INFO gateway.run: Session hygiene compression ... still streaming after 30s (last progress 0.0s ago) — extending wait (ceiling 600s)
14:20:01 INFO gateway.run: Session hygiene: 521 messages, ~175,873 tokens (actual) — auto-compressing ...
14:28:50 INFO gateway.run: Session hygiene: 521 messages, ~175,873 tokens (actual) — auto-compressing ...
14:28:12 INFO gateway.run: Demoting busy_input_mode 'interrupt' to 'queue' for session agent:main:feishu:dm:... because context compression is in flight
14:39:20 WARNING gateway.run: Session hygiene compression ... made no progress for 0.1s (total wait 629.5s, ceiling 600.0s); continuing without compression
```

Agent-side compression telemetry — three attempts, all aborted by the commit fence **with zero output**:

```text
[...]

PR #98424, opened 2026-08-30
fix(context): fail closed when preflight compression stalls

anhtahaylove:
## Summary
- mark owned compression attempts that end in a progress timeout
- fail turn-start preflight closed when an oversized payload remains unchanged after that timeout
- preserve existing structural no-op and direct-fence behavior

## Root cause
A stalled preflight compressor returned the original message list, so turn setup treated it like an ordinary no-op and continued into the provider call. The detached summary later lost its commit fence while the oversized provider request failed with context overflow.

## Tests
- `uv run --with pytest pytest tests/agent/test_preflight_compression_timeout_fail_closed.py tests/agent/test_compress_context_progress_timeout.py tests/agent/test_preflight_lock_defer.py -q` (19 passed)
- `uv run python -m py_compile run_agent.py agent/turn_context.py`
- `uv run --with ruff ruff check run_agent.py agent/turn_context.py tests/agent/test_preflight_compression_timeout_fail_closed.py tests/agent/test_compress_context_progress_timeout.py`
- `git diff --check`

### DIFF
--- a/agent/auxiliary_client.py
+++ b/agent/auxiliary_client.py
@@ -9535,6 +9535,7 @@ class _ChatStreamAccumulator:
         self._total_ceiling = total_ceiling
         self.content_parts: List[str] = []
         self.reasoning_parts: List[str] = []
+        self.reasoning_details: List[Any] = []
         self.tool_calls_acc: Dict[int, Dict[str, Any]] = {}
         self.finish_reason = None
         self.usage = None
@@ -9579,6 +9580,24 @@ class _ChatStreamAccumulator:
         if reasoning_piece and isinstance(reasoning_piece, str):
             self.reasoning_parts.append(reasoning_piece)
             made_progress = True
+        # OpenRouter-compatible reasoning models may stream their entire
+        # thinking phase through ``reasoning_details`` instead of
+        # ``reasoning`` / ``reasoning_content``.  Treat only details carrying
+        # actual text as forward progress: structural/signed envelopes must
+        # not keep an otherwise stalled compression alive indefinitely.
+        reasoning_details = getattr(delta, "reasoning_details", None)
+        if reasoning_details is None:
+            model_extra = getattr(delta, "model_extra", None)
+            if isinstance(model_extra, dict):
+                reasoning_details = model_extra.get("reasoning_details")
+        if isinstance(reasoning_details, list):
+            for detail in reasoning_details:
+                self.reasoning_details.append(detail)
+                if isinstance(detail, dict) and any(
+                    isinstance(detail.get(field), str) and detail[field]
+                    for field in ("summary", "thinking", "content", "text")
+                ):
+                    made_progress = True
         for tc in (getattr(delta, "tool_calls", None) or []):
             idx = getattr(tc, "index", 0) or 0
             acc = self.tool_calls_acc.setdefault(
@@ -9620,6 +9639,7 @@ class _ChatStreamAccumulator:
             content="".join(self.content_parts),
             tool_calls=tool_calls,
             reasoning="".join(self.reasoning_parts) or None,
+            reasoning_details=self.reasoning_details or None,
         )
         choice = SimpleNamespace(
             index=0,
--- a/agent/turn_context.py
+++ b/agent/turn_context.py
@@ -405,6 +405,21 @@ def compression_made_progress(
 _compression_made_progress = compression_made_progress
 
 
+class PreflightCompressionTimedOut(RuntimeError):
+    """Raised when an oversized turn cannot safely finish preflight."""
+
+
+def _fail_closed_after_preflight_timeout(agent, request_tokens: int) -> None:
+    """Stop an oversized turn instead of sending its unchanged provider payload."""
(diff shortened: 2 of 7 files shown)


liyangbing:
<p>This closes an important safety boundary: a progress-timeout result must not be treated as an ordinary compression no-op when the request is still oversized. The separate timeout marker also preserves the existing compression-lock defer path, which should not be turned into a false failure.</p>

<h3>Suggested contract</h3>

<ul>
<li>Guarantee at the turn boundary that <code>PreflightCompressionTimedOut</code> is converted into a bounded, user-visible recovery response and that no provider request is attempted afterward.</li>
<li>Keep the timeout marker scoped to one compression attempt and reset it on every new attempt, including manual <code>/compress</code>; a later successful compression must not inherit a stale fail-closed state.</li>
<li>Distinguish a progress timeout from a compression-lock defer, structural no-op, and an already-small payload in both logs and user-facing diagnostics.</li>
<li>Preserve the commit fence: a timed-out attempt must not expose a detached summary or partially committed transcript as if compression succeeded.</li>
<li>Make the retry contract explicit: tell the operator whether to run <code>/compress</code>, wait for another writer, or retry the original turn, and ensure the next retry can actually re-enter compression.</li>
<li>Keep sync and async turn paths equivalent, including exception translation, provider-call suppression, activity touching, and cleanup of per-attempt state.</li>
</ul>

<p>The new unit test proves the helper raises for the marker, but the strongest invariant is one level higher: a real <code>build_turn_context</code> or turn-start test should install a provider spy and assert that the unchanged oversized payload never reaches it. It should also assert that the exception is rendered through the normal recovery path rather than becoming an unclassified traceback.</p>

<h3>Regression matrix</h3>

<ul>
<li>progress timeout with unchanged message count and unchanged token estimate</li>
<li>progress timeout after a partial compression pass, with no detached-summary commit</li>
<li>compression-lock defer, structural no-op, and successful size-only compression</li>
<li>sync and async turn-start paths with a provider-call spy</li>
<li>manual <code>/compress</code> after an automatic timeout, followed by a retry of the original turn</li>
<li>provider unavailable, session missing, cancellation, and repeated timeout cleanup</li>
<li>user-visible recovery text, warning severity, activity timestamp, and persisted transcript state</li>
</ul>

<p>Evidence reviewed: <code>agent/turn_context.py</code>, <code>run_agent.py</code>, and <code>tests/agent/test_preflight_compression_timeout_fail_closed.py</code> in PR head <code>33a434d6395cf4e79e871b5806754a5c88570ff7</code>.</p>

anhtahaylove:
Thanks for the detailed contract review. I rebased this branch onto current `main` and added coverage for the boundary you called out:

- `build_turn_context` now converts the timeout marker into `PreflightCompressionTimedOut` only when the request remains oversized and compression made no row/token progress.
- `tests/agent/test_preflight_compression_timeout_fail_closed.py` installs a provider spy and proves the unchanged oversized payload is never sent; the timeout marker is reset per compression attempt, so a later manual/successful retry does not inherit stale fail-closed state.
- Compression-lock defer and ordinary structural no-op remain separate paths.
- The rebase preserves the newer total-ceiling/progress diagnostics, and `0889692520` also counts streamed `reasoning_details` as progress.

Verification after rebase:
- `uv run pytest -q tests/agent/test_preflight_compression_timeout_fail_closed.py tests/agent/test_compress_context_progress_timeout.py tests/agent/test_turn_context.py` — 39 passed.
- Focused reasoning-progress tests — 2 passed.

One pre-existing Windows timing assertion in `test_keepalive_chunks_do_not_reset_the_compression_fence` still observes exactly `0.0` elapsed immediately after fence creation; it is unrelated to this PR and the new/affected tests are green.

teknium1:
**Review from the P1 compression-wedge sweep (issues #98722 / #97948-B / #96953-adjacent).**

Verified the premise against current `main@26f178e5fa`: a stalled preflight compression still returns the input list, `_compression_made_progress` treats it as a structural no-op, and the unchanged oversized payload proceeds to the provider — `_last_compression_timed_out` / any typed timeout outcome does not exist on main (`grep` clean across `agent/` + `run_agent.py`). So the gap this PR closes is real and un-superseded by the Aug wave (`4f22543509` single-call lean, `087cc49a26` split-failure cooldown, `892f756b8c` worker teardown) — those reduce *how often* an attempt stalls; none of them prevents the doomed provider send after a stall.

CI is green on head `0889692520` and the branch is mergeable against current main; the rebase correctly preserves the newer total-ceiling diagnostics, and counting streamed `reasoning_details` as progress composes with the thinking-model behavior reported in #97963.

One coordination note for the maintainer: **#98741 overlaps this PR** — it introduces its own timeout marker (`_get_context_compression_timeout_state`, thread-local + per-agent lock) consumed at turn-start *and* the overflow-recovery loop. Landing both as-is would double-implement the marker. Since this PR is mergeable+green and #98741 is currently conflicting, the natural sequencing is: this PR carries the preflight fail-closed boundary; #98741 rebases on top and narrows to the overflow-recovery consumer + concurrency hardening of the shared marker. Merge decision stays with Teknium as always.

Issue #97898, opened 2026-08-29
Windows: agent filesystem traversal hydrates iCloud/OneDrive cloud-file placeholders


xngg1021:
## Environment
- Hermes Agent version: v0.20.6
- OS: Windows 11
- Cloud: iCloud for Windows (also applies to OneDrive Files On-Demand)

## Problem
On Windows, when the agent scans a directory tree (e.g. `find /c/Users/<user>` to
locate a `.git` directory), it stats every entry. Windows Cloud Files placeholders
(used by iCloud Drive / iCloud Photos and OneDrive Files On-Demand) hydrate on any
filesystem access — including command-line tools. Microsoft's own docs state:
"Whether you use file system APIs, the Command Prompt, or a desktop or a UWP app to
access a placeholder file, the file will hydrate."

In practice, this caused iCloudPhotos to download at ~53 MB/s while the agent was
only searching for a directory. The `find` process also survived the shell timeout
and kept traversing.

## Suggested improvement
On Windows, when traversing trees outside the working directory, skip or warn about
cloud-file sync roots (iCloud Drive, iCloud Photos, OneDrive). Better: detect the
cloud-placeholder attribute (FILE_ATTRIBUTE_RECALL_ON_DATA_ACCESS / CF_PLACEHOLDER)
and avoid stat-ing placeholder files during directory scans.

## Why this matters
A routine "find the repo" task turns into hundreds of MB (or GB) of unintended cloud
downloads — bandwidth and disk writes — and is easy to mistake for the cloud
client's own sync.

lEWFkRAD:
I'm picking this up — working on a fix with tests. Will open a PR shortly.

lEWFkRAD:
Submitted #98425 for the search-tool half: broad Windows searches now prune OneDrive/iCloud roots pre-traversal (rg globs, find -prune, grep fallback), with real-rg glob-semantics evidence — backslash globs never match an absolute Windows search root, so exclusion globs are absolute forward-slash spelled. The terminal half (find/du invoked directly by the agent) is intentionally untouched here and stays open on this issue.

MarxChangMRT:
## Additional Windows Desktop failure mode (cwd = `%USERPROFILE%`)

Confirming this on **Windows 11 + OneDrive Files On-Demand** with Hermes Desktop.

**Environment**
[...]

PR #98425, opened 2026-08-30
fix(search): prune OneDrive/iCloud sync roots from broad Windows file searches

lEWFkRAD:
## What

Broad `search_files` walks of a Windows profile (or any directory containing one) now **prune OneDrive and iCloud sync roots before traversal**, so ripgrep/find/grep never stat the dehydrated Cloud Files placeholders below them. Contributes to #97898 — the `terminal` `
**Why:** Windows Cloud Files (OneDrive Files On-Demand, iCloud for Windows) keep directory entries that hydrate on *first access* — and a plain `stat()` counts as access. A routine agent search of `C:\Users\<user>` therefore silently downloads the user's whole cloud library (the issue reporter measured ~53 MB/s of iCloud Photos from a "find the repo" scan). Prevention must happen before descent; post-filtering matches is too late — the bytes are already spent.

**How:** mirrors the existing macOS TCC precedent in the same file. New `_windows_cloud_search_exclusions()` resolves cloud roots from `%OneDrive%` / `%OneDriveConsumer%` / `%OneDrivePublic%` / `%iCloudDrive%` / `%I_CLOUD_DRIVE_LOCATION%` plus the default iCloud-for-Windows conventions (`~\iCloudDrive`, `~\iCloud Photos`, `~\Pictures\iCloud Photos`), and returns the ones contained in the search root. `_macos_search_exclusions()` becomes `_protected_search_exclusions()` (union of the two platform helpers, each a no-op off-platform) and every engine that already consumed it inherits the pruning for free: rg `--glob '!<root>/**'` (file + content search), `find -prune`, and the grep fallback's path-scoped prune. The `env.is_local` gate carries over — a Windows controller driving a Linux backend prunes nothing.

**Glob-spelling evidence (this is the part that was easy to get wrong):** verified against real ripgrep on Windows — `-g '!C:/Users/.../OneDrive/**'` (forward slashes, absolute, anchored at the search root) prunes; backslash spellings and profile-relative names **never match an absolute search root**, so exclusion globs are emitted forward-slash. A real-rg integration test asserts the descent actually stops (`@WINDOWS_ONLY`, skips elsewhere — forcing Windows-spelled globs on Linux CI would test the OS, not the code).

**Deliberate non-goals:**
- Searching *inside* a cloud root stays unpruned — one explicitly requested file may legitimately hydrate; walking the whole tree unattended is the bug.
- `tools/terminal_tool.py` (`find`/`du` run directly by the agent) is **not** touched — the terminal half of #97898 remains open and this PR says so on the issue.

## Tests

New `tests/tools/test_windows_cloud_placeholder_search.py` — 9 tests: unit (fake env + fake platform; no dependency on the CI box's OneDrive state), engine-command tests (rg file+content globs present, remote backend never prunes, warning surfaced), and the real-rg descent test.

- `pytest tests/tools/test_windows_cloud_placeholder_search.py tests/tools/test_macos_protected_search.py -q` → **18 passed, 2 failed**
- The 2 failures are **pre-existing on clean `origin/main` on this Windows box** (verified: same 2 fail unmodified) — they compare `str(Path)` backslash spellings against MSYS-converted command strings in the `find`/grep prune paths. Not macOS regressions; the rg-glob macOS tests all pass.
- `pytest tests/tools/test_file_operations*.py tests/tools/test_file_ops_cwd_tracking.py -q` → **9 failed, 83 passed — byte-identical tally to clean `origin/main`** on this box.
- `ruff check` on both changed files: clean.

### DIFF
--- a/tests/tools/test_windows_cloud_placeholder_search.py
+++ b/tests/tools/test_windows_cloud_placeholder_search.py
@@ -1,0 +1,166 @@
+"""Windows Cloud Files placeholder-safe broad searches (#97898).
+
+Unit tests are deterministic (fake env, fake platform). Real-rg integration
+tests skip on non-Windows: the exclusion globs are platform-correct by
+design, so forcing them cross-platform would test the OS, not the code.
+"""
+
+import shutil
+import sys
+from pathlib import Path
+
+import pytest
+
+import tools.file_operations as file_operations
+from tools.file_operations import ShellFileOperations, _windows_cloud_search_exclusions
+from tools.environments.local import LocalEnvironment
+
+class RecordingEnvironment:
+    def __init__(self, cwd):
+        self.cwd = str(cwd)
+        self.commands = []
+
+    def execute(self, command, cwd=None, **kwargs):
+        self.commands.append(command)
+        if command.startswith("test -e"):
+            return {"output": "exists\n", "returncode": 0}
+        if command.startswith("command -v"):
+            return {"output": "yes\n", "returncode": 0}
+        return {"output": "", "returncode": 1}
+
+
+def _fake_env(profile: Path) -> dict:
+    return {
+        "USERPROFILE": str(profile),
+        "OneDrive": str(profile / "OneDrive"),
+        "OneDriveConsumer": str(profile / "OneDriveConsumer"),
+        "iCloudDrive": str(profile / "iCloudDrive"),
+    }
+
+
+def _patch_windows(monkeypatch, profile: Path):
+    monkeypatch.setattr(file_operations, "_HOME", str(profile))
+    monkeypatch.setattr(file_operations.sys, "platform", "win32")
+    monkeypatch.setenv("USERPROFILE", str(profile))
+    for name in ("OneDrive", "OneDriveConsumer", "OneDrivePublic",
+                 "iCloudDrive", "I_CLOUD_DRIVE_LOCATION"):
+        monkeypatch.delenv(name, raising=False)
+    monkeypatch.setenv("OneDrive", str(profile / "OneDrive"))
+    monkeypatch.setenv("OneDriveConsumer", str(profile / "OneDriveConsumer"))
+    monkeypatch.setenv("iCloudDrive", str(profile / "iCloudDrive"))
+
+
+def test_broad_profile_search_prunes_cloud_roots(tmp_path):
+    profile = tmp_path / "Users" / "jeff"
+    exclusions = _windows_cloud_search_exclusions(
+        str(profile), cwd=str(tmp_path), env=_fake_env(profile), platform="win32"
+    )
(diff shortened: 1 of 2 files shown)


liyangbing:
<p>This is a valuable boundary fix: pruning OneDrive/iCloud roots before <code>rg</code>, <code>find</code>, or grep traversal is materially different from filtering matches after Cloud Files has already hydrated them. The explicit exception for a caller that starts inside a cloud root is also a clear and defensible contract.</p>

<h3>Suggested contract</h3>

<ul>
<li>Apply the same root-containment rule to every search engine and both targets (files and content); no engine should perform a preliminary <code>stat</code> or directory probe below an excluded root.</li>
<li>Keep the controller/backend boundary fail-safe: when <code>env.is_local</code> is false, do not derive exclusions from the controller's Windows profile or environment.</li>
<li>Make Windows path identity explicit for drive-letter case, slash spelling, trailing separators, UNC paths, and case-insensitive containment; do not let a sibling such as <code>OneDrive-old</code> match a protected root.</li>
<li>Deduplicate environment-derived and default iCloud roots, and ignore malformed, relative, or empty environment values without turning them into exclusions outside the user profile.</li>
<li>Keep the warning truthful: report which roots were skipped, distinguish cloud hydration from macOS privacy prompting, and do not claim a skip when the backend returned an error before traversal.</li>
<li>Document the symlink/junction boundary. If containment is lexical by design, say so and test that a reparse point cannot silently turn the broad search back into an unbounded cloud walk.</li>
</ul>

<p>The PR body reports two pre-existing Windows-box failures in the focused command tests and a byte-identical baseline in the broader file-operation tests. Please keep that distinction visible in CI and add a native Windows job for the real-ripgrep descent test; the non-Windows skip is correct for Windows-spelled glob behavior, but it otherwise leaves the most important hydration invariant unexecuted on typical CI.</p>

<h3>Regression matrix</h3>

<ul>
<li>OneDrive and iCloud roots discovered through each supported environment variable and each default profile location</li>
<li>profile search, explicit search at a cloud root, and nested search below a cloud root</li>
<li>rg file search, rg content search, <code>find -prune</code>, and grep fallback</li>
<li>local Windows, remote backend, non-Windows controller, and a missing or stale cloud root</li>
<li>drive-letter/case variants, UNC paths, sibling names, duplicate roots, junctions, and symlinks</li>
<li>an existing hydrated file beside a dehydrated placeholder, proving no descent before matching</li>
<li>permission denial or command failure, verifying the warning and error fields remain accurate</li>
</ul>

<p>Evidence reviewed: <code>tools/file_operations.py</code> and <code>tests/tools/test_windows_cloud_placeholder_search.py</code> in PR head <code>ceee223036bda9039713d5f4f763b5858f3029bb</code>.</p>

PR #97979, opened 2026-08-29
feat(tool-search): core-tool deferral — curated event-triggered set behind the bridge (−6.5K tok/call desktop, −49% schemas)

teknium1:
Maintainer-directed phase 2 of the context campaign (#95681): the first crossing of the old "core never defers" line. A curated set of 19 EVENT-TRIGGERED tools moves behind the tool_search/tool_describe/tool_call bridge by default. Opened for maintainer-run A/B testing before merge.

## The curated defer set (maintainer-specified)
computer_use · session_search · clarify · image_generate · todo_list · process_manage · cronjob_manage · and the full desktop GUI surface (drive_preview, gui_tour, desktop_preview, annotate_preview, show_tip, setup_mcp, desktop_project, close_terminal, apply_layout, read_terminal, read_window_below, focus_pane).

Design rationale: these are tools a session reaches for when something HAPPENS (user asks for a tour / cron job / screenshot), not the every-turn working set — a catalog stub is enough to find them. The working set (terminal, files, memory, browser_exec, delegation, skills, web, vision) stays eager, always.

## Measured (real desktop assembly, o200k_base, E2E)
- Before: 37 tools, 13,367 tok
- After: **18 tools, 6,883 tok (−6,484, −49%)**
- Desktop full context: ~18.2K → **~11.7K/call**; vs pre-campaign 31.7K ≈ **−63% all-in**
- tool_search schema grows 309→710 (the 19 catalog stub lines — the discovery surface), already netted in the above.

## Renames (ride along since deferral touches every registration)
todo→**todo_list**, cronjob→**cronjob_manage**, process→**process_manage**, tour→**gui_tour**, tip→**show_tip**. All reference layers swept (toolsets, dispatch branches in both executors, display/cute-messages, context_compressor summarizers, guardrails, turn_summary, coding_context, delegate_task deny-list, tui_gateway). **Legacy aliases at every dispatch seam** (`_LEGACY_TOOL_ALIASES` in model_tools + both executors): old sessions replaying `todo`/`cronjob`/... keep working; new schemas only advertise new names. E2E-verified both routes (bridge call with new name; direct call with old name).

## Mechanism
- `_DEFAULT_DEFERRED_TOOLS` in tools/tool_search.py; `ToolSearchConfig.defer_tools` + `effective_defer_tools`.
[...]

PR #97930, opened 2026-08-29
feat(system_prompt): two-line conversation clock — anchored start + rebuild-day line (salvages #96224)

teknium1:
Resolves the 'Conversation started drifts out of date' complaint (kxee's report, maintainer design). Salvages the anchor from #96224 by cherry-pick so @bobaba76's authorship survives, then builds the maintainer's two-line design on top — #96224 alone was rejected because a FROZEN start date creates the opposite problem in forever-sessions (Bot Mode, messenger channels): the model believes it still lives on the session's birth day.

## Design (maintainer-specified)
- **Line 1 (stable):** `Conversation started: <date>` — anchored to the REAL session start via #96224's ladder (session_id-embedded YYYYMMDD_HHMMSS → agent.session_start → build time). Byte-stable across every rebuild.
- **Line 2 (refreshing, multi-day sessions only):** `Today's date (as of the last context rebuild): <date> — trust this over the start date for what day it is now; query tools for exact time.`

## Cache accounting
The second line only changes when `now` crosses a day boundary AND the prompt is rebuilt — and rebuilds happen exactly at compaction/fresh-agent boundaries where the prefix cache is already invalidated. Same-day sessions render the single line, byte-identical for the day. Net: zero additional cache churn over the status quo.

## Precedence-teaching
The line says explicitly which date wins ("trust this over the start date") — without it, two dates in the prompt is a coin-flip. Timeless Bot Chat prompts (dateless by design) are untouched — verified by test.

## Salvage notes
Both #96224 commits cherry-picked intact (fix + tests). One salvaged test made TZ-portable (its 23:05-local timestamp read as the next day when converted to UTC on CI boxes east of the stamp; moved to midday — this is also what closed #96224's original CI failure). New tests: multi-day session gets the second line; same-day session keeps the one-line shape; timeless Bot Chat unaffected. 44/44 system-prompt tests green.

PR #72253, opened 2026-07-26
refactor(agent): consolidate SOUL.md identity resolution and pin drift propagation

Sora-bluesky:
Editing SOUL.md never reached a continuing session: the restore path reuses the persisted system prompt verbatim, and `_stored_prompt_matches_runtime` only rejects Model/Provider/cwd/Platform drift, not identity content drift (#68563).

This PR consolidates identity resolution into `resolve_identity_block()`, now the single source that the prompt builder uses, and pins SOUL.md drift propagation with regression tests. The compaction-gate staleness check earlier versions of this PR added was superseded by #98426's unconditional rebuild at the compaction boundary: a drifted identity now simply makes the fresh build differ from the cached bytes, and the tests here pin that propagation. `stored_identity_is_stale` stays as a tested helper with no production caller today. Restore-path I/O is unchanged (AGENTS.md only exempts context compression from the no-mid-conversation-rewrite rule).

PR #98426, opened 2026-08-30
feat(compaction): system prompt always rebuilds at the commit boundary — updates finally reach long-lived sessions

teknium1:
Maintainer-directed (#95681 arc). Companion to the compaction tool-refresh: tools already rebuild from the live registry at the compaction commit, but the system prompt was restored byte-for-byte from the session snapshot FOREVER — so prompt-side improvements (guidance diets, new blocks like the two-line clock, renames like #97979's) never reached a long-lived session. Bot Mode forever-chats and gateway channels would carry their birth prompt for months while their tools array moved on — guidance drifting against live reality.

## Maintainer design decisions (verbatim intent)
- "after an update, we have to rebuild the system prompt & tools... I don't want an update detector though" — no detector anywhere: the rebuild IS the change detector, because it runs the live builder and compares bytes.
- "Why would we not update the system prompt at compaction?" — no fingerprint gate either. ALWAYS rebuild at the commit boundary; the prefix cache is already dead there by definition, so the rebuild is free.
- Plugin sections too: the conservative freeze-and-splice idea was rejected — a plugin section is just another prompt block carrying state; freezing it while memory/skills/guidance refresh recreates the stale-block disease inside plugin-land.

## Semantics
1. Commit boundary runs the LIVE builder every time.
2. Output byte-equal to the stored prompt — keep the ORIGINAL string object (object identity preserved for KV/prefix caches keyed on it; reconstruct_static_prefix path unchanged). The common case costs one string compare.
3. Output differs — rebuilt prompt wins, drift logged ("Compaction rebuilt a drifted system prompt (session=..., N -> M chars)").
4. Plugin sections: invalidate_system_prompt clears the frozen snapshot (stashing previous bytes); the rebuild re-renders via the same render path as session start. A plugin whose render RAISES falls back to its last good section — fail-open guard, not a freeze. Restore paths between compactions keep today's byte-stable behavior untouched.
5. The old keep-prompt gate (_cached_prompt_reflects_builtin_memory containment) is retired from the commit site — byte equality of the full rebuilt output subsumes it exactly.

## What this changes for users
Long-lived sessions converge on the current prompt at their next compaction instead of never: memory saved mid-life finally appears (the two-line clock's "as of the last context rebuild" now labels exactly when), dieted guidance replaces stale text, and future renames stop presenting old-name guidance against new-name tools. No mid-conversation cache break is introduced anywhere — the ONLY rebuild point is the already-sanctioned one.

## Verification
New contract suite (tests/test_compaction_prompt_rebuild.py, 6 tests): byte-equality keep-prompt pin, containment-gate-must-not-return pin, drift logging, plugin snapshot stash/clear, fail-open to previous bytes, empty fallback without previous. Tool-refresh companion test updated to the new commit-site shape. 44 system-prompt tests green. The e2e that matters (forever-session picks up a builder change at compaction) is source-pinned here; live confirmation is one /compress away after merge.

### DIFF
--- a/agent/conversation_compression.py
+++ b/agent/conversation_compression.py
@@ -3979,29 +3979,21 @@ def compress_context(
                 exc_info=True,
             )
 
-        # Built-in memory is the only system-prompt input that a normal
-        # compaction reloads. When the cached prompt already embeds the
-        # freshly-reloaded memory blocks verbatim, keep the exact cached
-        # prompt so local backends retain their KV-cache prefix. Containment
-        # (not before/after snapshot equality) is required: fresh-agent
-        # surfaces restore the cached prompt from the session DB, where it
-        # can predate mid-session memory writes the in-memory snapshot has
-        # already absorbed. External providers can change their own prompt
-        # block during on_pre_compress(), so they retain the rebuild path.
-        if (
-            cached_system_prompt is not None
-            and getattr(agent, "_memory_manager", None) is None
-            and _cached_prompt_reflects_builtin_memory(agent, cached_system_prompt)
-        ):
+        # ALWAYS rebuild the prompt at the admitted-commit boundary
+        # (maintainer-directed, #95681 arc). The previous "keep-prompt"
+        # containment branch put the OLD bytes back whenever the reloaded
+        # memory blocks were already embedded — which meant prompt-builder
+        # changes (guidance diets, new blocks, renames) NEVER reached a
+        # long-lived session. The cache argument for keeping bytes was
+        # hollow: when nothing changed, the rebuild is byte-identical and
+        # local KV prefixes survive on equality; when something changed,
+        # the cache was stale by definition and propagation is the point.
+        # Preserve OBJECT identity on byte-equality for backends that key
+        # on it.
+        rebuilt_system_prompt = agent._build_system_prompt(system_message)
+        if cached_system_prompt is not None and rebuilt_system_prompt == cached_system_prompt:
             new_system_prompt = cached_system_prompt
             agent._cached_system_prompt = cached_system_prompt
-            # _invalidate_system_prompt() above also cleared the
-            # cross-session-stable prefix marker boundary. The kept prompt
-            # is byte-identical, so reconstruct the stable tier and reuse
-            # it ONLY when the kept prompt still literally starts with it
-            # (same startswith gate as the restore path); otherwise the
-            # request layer falls back to the legacy single-breakpoint
-            # layout with the prompt bytes untouched.
             from agent.system_prompt import reconstruct_static_prefix
 
             reconstruct_static_prefix(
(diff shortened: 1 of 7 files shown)


liyangbing:
<p>This is a strong fix for long-lived sessions: rebuilding at the admitted compaction commit boundary lets prompt-builder changes converge without introducing a new update detector or a mid-turn cache break. Keeping the original string object only on byte equality is a precise optimization boundary.</p>

<h3>Suggested contract</h3>

<ul>
<li>Make the commit ordering explicit: the live builder must run after the intended memory/tool state is visible, and the rebuilt prompt must be the one persisted and used for the post-compaction request.</li>
<li>Keep prompt changes out of an active provider call. A compaction that is deferred, cancelled, or fails its commit fence must not partially replace the session's prompt or plugin sections.</li>
<li>Scope the plugin fallback bytes to the same plugin identity/configuration and rebuild generation. A removed or replaced plugin must not resurrect an old section merely because a later renderer raises.</li>
<li>Preserve the byte-equality fast path and static-prefix identity guarantees, but add an invariant for the changed-prompt path: rebuilt static-prefix metadata must correspond to the new bytes and must not retain an old prefix marker.</li>
<li>Keep plugin render failures observable and bounded. A last-good section may be retained, but the report/log should identify the plugin and distinguish fallback from a successful re-render.</li>
<li>Ensure repeated compactions are idempotent: equal output reuses the original object, changed output becomes the new baseline, and a later failure falls back only to the correct last-good generation.</li>
</ul>

<p>The source-level tests pin the always-build and equality gate, and the plugin unit tests pin the fallback helper. The highest-value missing witness is an end-to-end long-lived Session test that changes a builder input, compacts, then asserts the persisted/requested prompt changed while tools and memory remain aligned; it should also cover plugin removal and a failed re-render after one successful refresh.</p>

<h3>Regression matrix</h3>

<ul>
<li>byte-identical rebuild with original string identity and reconstructed static prefix</li>
<li>changed built-in memory, guidance, skill, and plugin prompt blocks at compaction</li>
<li>deferred, cancelled, failed, and successful compaction commit boundaries</li>
<li>plugin render success, failure with last-good bytes, failure without previous bytes, removal, replacement, and configuration change</li>
<li>repeated compactions with alternating equal, changed, and failing builder output</li>
<li>long-lived gateway/Bot Mode session persistence and next-request prompt/tool alignment</li>
<li>sync/async or local/remote prompt consumers, if both use this commit path</li>
</ul>

<p>Evidence reviewed: <code>agent/conversation_compression.py</code>, <code>agent/system_prompt.py</code>, <code>tests/test_compaction_prompt_rebuild.py</code>, and <code>tests/test_compaction_tool_refresh.py</code> in PR head <code>5ea109b3234c60c237c37c40206d279cc05bc65c</code>.</p>

teknium1:
Second commit (maintainer request): the always-rebuild made one latent gap in the #97930 ladder matter — compaction ROTATES session ids, and each rotated id embeds its own mint time, so after enough compactions rung 1 would quietly re-birth the conversation at its latest rotation. New rung 0: walk to the lineage root (same get_conversation_root walk Portal usage tagging uses) and prefer ITS embedded stamp — the conversation's true birth survives every compaction. Fail-open at every step: walk raises → segment id; root has no embedded stamp (legacy/imported lineage) → segment id; no DB → exactly the old ladder. 4 new ladder tests; 53 green across system-prompt + rebuild suites.

Sora-bluesky:
Cross-reference: open PR #72253 (SOUL.md identity drift) also edits the compaction keep-prompt path this PR replaces. A trial merge in both directions conflicts in `agent/conversation_compression.py`, and the two PRs' changed files intersect on `agent/conversation_compression.py` and `agent/system_prompt.py`. Flagging the overlap early. Once this lands I'll rebase #72253 onto it.

PR #98429, opened 2026-08-30
fix(tools): stop terminal guards from blocking benign commands containing '&'

albert748:
Terminal guards scanned raw command text without respecting what actually executes, hard-blocking ordinary agent workflows (exit_code=-1, status 'error') on command text that is never run as shell syntax. This is the complete fix set for the reported false-positive families, collapsed into one commit against the current main:

1. **Heredoc bodies are inert data for their interpreter, not shell syntax.** Quoted (inert) heredoc bodies after `cd <path> &&` / `source <path> &&` prologue links, `VAR=...` assignments, `env`, `timeout <N>` wrappers and `~`/relative interpreter paths are now masked; quoted paths or command substitutions stay fail-closed, and bash/sh consumers keep their bodies scanned.

2. **Compound openers also mask inert bodies** (`python3 - <<'EOF' 2>&1 | grep -v X`, `build && python3 - <<'EOF'`) — a bitwise `&` in interpreter program text is no longer reported as shell backgrounding. An opener with a *shell* consumer on the line (`cat <<'EOF' | bash`) stays fail-closed.

3. **A grep word inside a double-quoted command substitution** made the operand span unprovable and the whole command failed the hardline check as "malformed executable payload" (not bypassable even with --yolo). Only that grep word is skipped now; the raw command still reaches every other detector.

4. **Long-lived detection is scoped to the command's own start words**, so `ps aux | grep uvicorn` / `journalctl -u X | grep uvicorn` (process inspection) are not mistaken for `python -m uvicorn` (server start). Bare `docker compose up` stays flagged, but `docker compose up -d|--detach` returns immediately (daemon-owned) and is not long-lived — this unblocks `compose down && compose up -d` restart workflows.

5. **Bounded executions are not long-lived.** A `timeout <N>` wrapper bounds the process (smoke/restart tests), so `timeout 15 .venv/bin/python -m uvicorn ...` passes; unbounded starts are still flagged.

6. **`vite build` is a one-shot build**, not a dev server — excluded from the long-lived pattern, while bare `vite` / `vite dev` stay flagged. `vite` was also missing from the long-lived start-word set, so bare dev servers actually slipped past detection; that gap is closed too.

Tests: new cases for prologue-link heredocs, compound openers (python masked / shell fail-closed), grep-substitution non-malformed, inspection-vs-start scoping, detached compose, bounded `timeout` starts and `vite build`. All existing tests pass (guards/heredoc/hardline/approval/terminal suites).

### DIFF
--- a/tests/tools/test_blocked_command_guidance.py
+++ b/tests/tools/test_blocked_command_guidance.py
@@ -82,3 +82,72 @@ class TestBackgroundGuidanceRecipes:
 
     def test_quoted_ampersand_not_flagged(self):
         assert _foreground_background_guidance('git commit -m "a & b"') is None
+
+    def test_inspection_of_uvicorn_is_not_server_start(self):
+        """`ps aux | grep uvicorn` inspects; the server word is only a grep
+        operand, so the command must not be nudged to background mode."""
+        assert _foreground_background_guidance("ps aux | grep uvicorn | grep -v grep | head -3") is None
+        assert _foreground_background_guidance(
+            "journalctl --user -u app-server -n 5 | grep uvicorn"
+        ) is None
+        assert _foreground_background_guidance(
+            "kill 4158132 && sleep 2 && ps aux | grep uvicorn | grep -v grep | wc -l"
+        ) is None
+
+    def test_real_python_uvicorn_start_still_flagged(self):
+        assert _foreground_background_guidance(
+            "cd ~/proj && .venv/bin/python -m uvicorn src.server.main:app --port 8000"
+        ) is not None
+        # `timeout <N>` bounds the process — a smoke/restart test, not a
+        # long-lived foreground server.
+        assert _foreground_background_guidance(
+            "cd ~/proj && timeout 15 .venv/bin/python -m uvicorn src.server.main:app --port 8001"
+        ) is None
+
+    def test_timeout_bounded_server_start_not_flagged(self):
+        """A `timeout <N>` wrapper makes the command bounded, so even a server
+        start cannot occupy the foreground forever."""
+        assert _foreground_background_guidance(
+            "systemctl --user stop app-webui.service 2>/dev/null; sleep 2; cd ~/proj && timeout 15 .venv/bin/python -m uvicorn src.server.main:app --host 0.0.0.0 --port 8001"
+        ) is None
+        assert _foreground_background_guidance(
+            "cd ~/proj && MOCK=1 timeout 8 .venv/bin/python -m uvicorn src.server.main:app --host 127.0.0.1 --port 8011"
+        ) is None
+        assert _foreground_background_guidance("timeout 60s npm run dev") is None
+        # unbounded still flagged
+        assert _foreground_background_guidance("npm run dev") is not None
+
+    def test_vite_build_is_not_server_start(self):
+        """`vite build` is a one-shot build, not a long-lived dev server.
+        Bare `vite` / `vite dev` are dev servers and stay flagged."""
+        assert _foreground_background_guidance(
+            "cd ~/proj/web && npx vite build 2>&1"
+        ) is None
+        assert _foreground_background_guidance("npx vite build") is None
+        assert _foreground_background_guidance("vite") is not None
+        assert _foreground_background_guidance(
+            "cd ~/proj && npx vite dev"
+        ) is not None
+
+    def test_npm_docker_starts_still_flagged(self):
+        assert _foreground_background_guidance("cd ~/proj && npm run dev") is not None
+        assert _foreground_background_guidance("docker compose up") is not None
+
+    def test_detached_docker_compose_up_is_not_server_start(self):
+        """`docker compose up -d|--detach` returns immediately (daemon-owned),
+        so the command line itself is NOT a long-lived foreground process."""
+        assert _foreground_background_guidance("docker compose up -d") is None
+        assert _foreground_background_guidance("docker compose up --detach") is None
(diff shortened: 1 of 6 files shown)


liyangbing:
<p>This is a good guard-precision fix: it distinguishes shell syntax that actually executes from inert Python heredoc content, keeps genuinely unprovable prologues fail-closed, and scopes long-lived detection to the command's starting executable so process inspection is not mistaken for a server launch.</p>

<h3>Suggested contract</h3>

<ul>
<li>Keep the security invariant explicit for the grep change: skipping an unprovable operand must never skip the raw command from dangerous-command scanning, and genuinely malformed shell quoting/options must still hardline. Add a test that combines the quoted substitution with a dangerous command outside the grep operand.</li>
<li>Document the trust boundary of a plain <code>source path &amp;&amp; python heredoc</code>. Sourcing a file executes arbitrary shell code before the interpreter; masking only the heredoc body is safe only if the guard still scans the prologue and the policy intentionally does not inspect the sourced file's contents.</li>
<li>Ensure heredoc masking is consumer-specific and wrapper-complete. Test <code>env</code>, <code>sudo</code>, <code>timeout</code>, <code>nohup</code>, <code>setsid</code>, <code>uv run</code>, absolute/tilde/relative interpreter paths, pipelines, substitutions, and mixed <code>;</code>/<code>&amp;&amp;</code>/<code>||</code> separators for both false positives and false negatives.</li>
<li>Make long-lived detection parse shell structure rather than relying on a quote-naive segment split wherever command substitutions or process substitutions can contain <code>|</code>, <code>&amp;&amp;</code>, or service-looking words. A detector should not miss a real start nested in a supported wrapper, nor flag a nested inspection as a start.</li>
<li>Keep the user contract consistent across approval tiers: benign quoted <code>&amp;</code> should proceed, real backgrounding and long-lived foreground starts should retain their existing guidance, and malformed/ambiguous input should retain the stricter result.</li>
<li>Preserve diagnostics that explain which guard matched and which command span was treated as inert; this is important when an operator sees a previously blocked command become allowed.</li>
</ul>

<p>The new tests cover the reported three families, but a direct metamorphic pair would strengthen the change: prepend an inspection pipeline such as <code>ps | grep uvicorn</code> and assert the server-start verdict is unchanged for the actual start command, then vary only the heredoc consumer or prologue and assert the body-mask decision changes as intended.</p>

<h3>Regression matrix</h3>

<ul>
<li>Python/pandas bitwise <code>&amp;</code> inside quoted heredocs with each supported prologue and wrapper</li>
<li>bash/sh consumers, quoted/substitution paths, missing separators, malformed quoting, and executable payloads that must remain visible</li>
<li>grep inside double-quoted command substitution, genuinely malformed grep options, dangerous text outside the operand, and nested substitutions</li>
<li>real starts for Python/uvicorn, npm/pnpm/yarn/bun, Docker Compose, Vite, Next, Gunicorn, and Nodemon</li>
<li>inspection commands using <code>ps</code>, <code>journalctl</code>, <code>grep</code>, <code>curl</code>, pipelines, and <code>grep -v</code> variants</li>
<li>wrappers, env assignments, <code>cd</code>/<code>source</code> links, separators, command substitutions, process substitutions, and quoted arguments</li>
<li>approval output, hardline guidance, background guidance, and fail-closed behavior for ambiguous syntax</li>
</ul>

<p>Evidence reviewed: <code>tools/approval.py</code>, <code>tools/shell_heredoc.py</code>, <code>tools/terminal_tool.py</code>, and the focused guard tests in PR head <code>b76c1d1375059021168b55b3c7ab7dfecee6b31c</code>.</p>

albert748:
Update summary — force-push `b76c1d1375` → `55eeaedfc3`: single commit rebased on current main.

**What changed on top of the original submission**
- **docker**: `docker compose up -d|--detach` is daemon-owned and returns immediately — no longer flagged as a long-lived server start; bare `docker compose up` stays flagged. This unblocks `compose down && compose up -d` restart workflows (described in point 4 of the updated description).
- **Bounded runs**: a `timeout <N>` wrapper makes the process bounded, so smoke/restart tests (`timeout 15 .venv/bin/python -m uvicorn ...`) are no longer hard-blocked; unbounded starts are still flagged (point 5).
- **vite**: `vite build` is a one-shot build and no longer flagged; bare `vite` / `vite dev` stay flagged — and `vite` was added to the long-lived start-word set (it was previously missing, so bare dev servers actually slipped past detection) (point 6).
- **Compound heredoc openers**: `python3 - <<'EOF' 2>&1 | grep -v X` and `build && python3 - <<'EOF'` now mask the inert interpreter body, so a bitwise `&` is no longer reported as shell backgrounding; openers with a shell consumer on the line (`cat <<'EOF' | bash`) stay fail-closed (point 2).

**Checks performed**
- Test suites: blocked-command guidance, heredoc background guard, hardline blocklist, approval, terminal-tool — all pass (433 tests, `pytest`; guard regression suite included).
- Regression audit: every terminal-guard error in the session store over the audit window (443 errors → 272 guard-related → 201 unique commands) was re-evaluated against the fixed guards. All false-positive families now pass; genuine backgrounding (`cmd &`), shell wrappers (`nohup`), bare `docker compose up`, unbounded server starts and hardline destroys remain blocked.
- Lint/CI: no new Pyright diagnostics; repository hooks clean.
- Neutralization: no personal paths, machine or project-specific tokens in code, tests, commit message or PR description. Real command samples that would reveal contributor environment details were intentionally excluded from the PR (kept local only).

The PR stays a single commit based on current main, ready for review.

PR #93989, opened 2026-08-24
fix(models): strip foreign vendor slugs on native providers

liuhao1024:
## What does this PR do?

`normalize_model_for_provider` was asymmetric for aggregator-namespaced ids (#93980): `glm-5.2` gains the `z-ai/` prefix when normalized for OpenRouter, but switching the provider afterwards handed the prefixed form verbatim to a native API — `ollama-cloud` only serves the bare id, so every `goal_judge` auxiliary call 404'd with `model "z-ai/glm-5.2" not found`. The judge retried the bad model every turn, and after `DEFAULT_MAX_CONSECUTIVE_TRANSPORT_FAILURES` consecutive failures the goal auto-paused pointing at the `goal_judge` config — which was fine; only the model id spelling differed between providers. This makes the `_MATCHING_PREFIX_STRIP_PROVIDERS` branch also strip a prefix that is a **known vendor slug** (the values of `_VENDOR_PREFIXES`), so the native provider receives the bare id it actually serves.

## Related Issue

Fixes #93980

## Type of Change

- [x] 🐛 Bug fix

## Changes Made

- `hermes_cli/model_normalize.py`:
  - new `_KNOWN_VENDOR_SLUGS` — the values of `_VENDOR_PREFIXES` (the vendor slugs the aggregator namespacing can stamp onto an id, e.g. `z-ai`).
  - the `_MATCHING_PREFIX_STRIP_PROVIDERS` branch: after the existing matching-prefix strip, if the result still carries a slash and the prefix is a known vendor slug, strip it. Other slash-bearing names pass through untouched, so a proxy-required routing prefix (LiteLLM's `ollama/glm-5.2` on a custom endpoint — the case `_strip_matching_provider_prefix`'s docstring explicitly guards) is never mangled. The aggregator direction is unchanged: bare `glm-5.2` still becomes `z-ai/glm-5.2` on OpenRouter.
- `tests/hermes_cli/test_model_normalize.py` — new `TestForeignVendorSlugStrippedOnNativeProviders` (5 tests): the reporter's exact shape (`z-ai/glm-5.2` → `glm-5.2` on ollama-cloud); other matching-strip providers (`openai/gpt-5.4` on zai, `anthropic/…` on gemini); bare and matching-prefix forms unchanged; the non-vendor proxy prefix (`ollama/glm-5.2` on custom) passes through; the OpenRouter namespacing direction untouched.

## How to Test

[...]

PR #93983, opened 2026-08-24
fix(model_normalize): strip foreign vendor prefix on bare-name providers (#93980)

startakovsky:
Fixes #93980.

## Problem

When the active model is the OpenRouter spelling of GLM (`z-ai/glm-5.2`) but the
provider resolves to `ollama-cloud`, every call — the main loop *and* the goal
judge — 404s with `model "z-ai/glm-5.2" not found`. `ollama-cloud` is in
`_MATCHING_PREFIX_STRIP_PROVIDERS`, and that branch only calls
`_strip_matching_provider_prefix(name, provider)`, which strips a prefix **only
when it normalizes equal to the target provider**. A foreign `z-ai/` prefix is
therefore left intact and reaches ollama.com verbatim, which serves the bare
`glm-5.2`. Because `auxiliary.goal_judge` runs with `provider: auto` it inherits
the mismatched id, the `NotFoundError` is classified as a transport failure, and
after N consecutive failures the goal auto-pauses pointing at the `goal_judge`
config — which is not the problem. 15 logged failures in 3 bursts.

## Approach

Adds a new `_FOREIGN_VENDOR_STRIP_PROVIDERS` frozenset and, in the
`_MATCHING_PREFIX_STRIP_PROVIDERS` branch, strips any leading `vendor/` segment
for members of that set before the existing matching-prefix repair runs. The
`_LOWERCASE_MODEL_PROVIDERS` lowercasing still runs after the strip, unchanged.

**The set contains exactly one provider: `ollama-cloud`.** This is deliberately
scoped rather than a blanket any-vendor strip across the whole
`_MATCHING_PREFIX_STRIP_PROVIDERS` set, because a blanket rule could regress a
provider that legitimately accepts vendor-prefixed ids or uses a prefix for
routing. The other members (`zai`, `kimi-coding`, `kimi-coding-cn`, `minimax`,
`minimax-oauth`, `minimax-cn`, `alibaba`, `qwen-oauth`, `xiaomi`, `arcee`,
`custom`, `gemini`, `xai`) keep their current behavior exactly.

`custom` is **excluded on purpose** and must stay excluded:
`_strip_matching_provider_prefix` deliberately preserves prefixes there, since
`ollama/glm-5.2` may be the actual routing prefix a LiteLLM-style proxy in front
of a custom endpoint requires. There is an explicit regression test for this.

New behavior:

```python
normalize_model_for_provider('z-ai/glm-5.2', 'ollama-cloud')  # -> 'glm-5.2'
[...]

PR #71558, opened 2026-07-25
fix(cli): centralize matching provider: and provider/ prefix stripping

briandevans:
## What does this PR do?

`hermes chat -m openai-codex:gpt-5.6-sol` sends the redundant prefix straight to the provider API. `_strip_matching_provider_prefix()` — the alias-aware helper every native-provider branch shares — split on `/` only, so the equivalent `provider:model` shorthand was invisible to it.

This centralizes both separators in that one helper, keeping its match guard, and picks up every provider branch that routes through it — **including the one branch that did not route through it at all**, the catalogue-backed NVIDIA repair (see "The fifth branch" below).

The reachable chain on current `main` — note `parse_model_input()` is **not** on it (its only non-test caller is `acp_adapter/server.py:831`):

- `cli.py:4039` stores `--model` verbatim; only a `moa:` prefix is special-cased, on the next line.
- `hermes_cli/cli_agent_setup_mixin.py:172` → `cli.py:5625` hands that un-split string to `normalize_model_for_provider()`.
- The colon prefix survives normalization and reaches the API.

The same input hits the same bug from the Desktop/dashboard model switch (`hermes_cli/web_server.py:1532`), the `/model` command (`hermes_cli/model_switch.py:1485`), and `model.default` in `config.yaml`.

> Rebased onto current `main` (`70c6cf8e7ef`). All numbers below are re-measured against that base.

## The fifth branch — catalogue-backed NVIDIA repair

`normalize_model_for_provider()` dispatches to **six** provider branches that repair a redundant provider prefix. Five of them now call `_strip_matching_provider_prefix()` first. The sixth — `_CATALOGUE_PREFIX_REPAIR_PROVIDERS` (currently `nvidia`), added by #78856 — never stripped a prefix at all:

```python
if provider in _CATALOGUE_PREFIX_REPAIR_PROVIDERS:
    return _repair_prefix_from_catalogue(name, provider)   # no strip, ever
```

`_repair_prefix_from_catalogue()` returns early whenever the id contains a `/` — the guard that keeps a genuine `nvidia/nemotron-…` slug **and** a self-hosted `nvidia/my-local-nim` addressable — and otherwise compares the *whole* id against each catalogue entry's post-slash suffix. A colon-prefixed id carries no slash, so it cleared the short-circuit, then matched no suffix, and passed through untouched:

```python
>>> normalize_model_for_provider("nvidia:nemotron-3-ultra-550b-a55b", "nvidia")
[...]

PR #34015, opened 2026-05-28
fix: strip provider prefix before passing model name to aggregator APIs

dangelo352:
## Problem

When a model is configured with a Hermes provider prefix (e.g. `openrouter/moonshotai/kimi-k2.5`), `normalize_model_for_provider` passes the full string — including the `openrouter/` prefix — to `_prepend_vendor()`. Since `_prepend_vendor` sees a `/` in the string, it returns it as-is.

**Result:** OpenRouter receives `openrouter/moonshotai/kimi-k2.5` as the model ID and rejects it with: `openrouter/moonshotai/kimi-k2.5 is not a valid model ID`. This affects all users selecting models through the instance webchat GUI.

## Fix

Add `_strip_matching_provider_prefix(name, provider)` before `_prepend_vendor(name)` in the aggregator branch (5 lines). This strips the `openrouter/` (or `nous/`, `ai-gateway/`, `kilocode/`) prefix so only the `vendor/model` slug reaches the aggregator.

Other provider branches (`_MATCHING_PREFIX_STRIP_PROVIDERS` etc.) already do this — the aggregator path was simply missing it.

### Before
`openrouter/moonshotai/kimi-k2.5` → sent as-is → ❌ rejected

### After
`openrouter/moonshotai/kimi-k2.5` → stripped to `moonshotai/kimi-k2.5` → ✅ accepted

PR #98435, opened 2026-08-30
fix(model_normalize): strip matching lmstudio/ prefix

Sinojen:
## What does this PR do?

`lmstudio` is missing from every normalization set in `hermes_cli/model_normalize.py`, so a `lmstudio/<id>` model name reaches LM Studio's native server verbatim. LM Studio validates the model id against the ids it has downloaded and rejects the prefixed form with a **non-retryable** HTTP 400, so every turn dies until the prefix is removed by hand.

Observed in production (`agent.conversation_loop`, provider `lmstudio`, `base_url=http://127.0.0.1:1234/v1`):

```
ERROR Non-retryable client error: Error code: 400 - {'error': {'message':
'Invalid model identifier "lmstudio/huihui-qwen3.8-27b-abliterated".
Please specify a valid downloaded model (e.g., huihui-qwen3.8-27b-abliterated,
qwen3.8-4b, ...)', ... 'code': 'model_not_found'}}
```

This looks like a plain omission rather than a deliberate exclusion: `ollama-cloud` — the analogous local-server provider — is already in `_MATCHING_PREFIX_STRIP_PROVIDERS`, and the set's own docstring describes this exact case ("Direct providers that accept bare native names but should repair a matching `provider/` prefix when users copy the aggregator form into config.yaml").

The prefixed form is not exotic: the repo's own test suite drives it through `switch_model` in `tests/run_agent/test_switch_model_context.py` (`assert model == "lmstudio/new-model"` with provider `lmstudio`).

Reproduced against current `main` (`2635035`):

```python
>>> from hermes_cli.model_normalize import normalize_model_for_provider as n
>>> n("lmstudio/qwen3-27b", "lmstudio")
'lmstudio/qwen3-27b'          # prefix survives -> HTTP 400 every turn
>>> n("ollama-cloud/qwen3", "ollama-cloud")
'qwen3'                       # analogous provider is handled
```

`_MATCHING_PREFIX_STRIP_PROVIDERS` is the correct bucket rather than a broader one: it strips only a prefix that *matches the provider*, so `lmstudio/x` → `x` while a non-matching prefix (e.g. a proxied `openai/x` on an endpoint fronted by LM Studio) is left untouched.

## Related Issue

No existing issue — searched open and closed issues and PRs for `lmstudio prefix`, `model_normalize`, `normalize_model_for_provider`, `_MATCHING_PREFIX_STRIP_PROVIDERS`, `Invalid model identifier`, and `model_not_found`. The nearest hits are unrelated: #54567 (merged) normalizes LM Studio *base URLs*, not model ids, and #93983 / #93989 / #71558 / #34015 touch the same file but neither add `lmstudio` nor make the strip generic (#93989 strips *foreign* vendor slugs, and only for providers already in this set).

## Type of Change

## Changes Made

- `hermes_cli/model_normalize.py` — add `"lmstudio"` to `_MATCHING_PREFIX_STRIP_PROVIDERS`, with a comment recording the API error the omission produces.
- `tests/hermes_cli/test_model_normalize.py` — add `TestLMStudioMatchingPrefixStrip`: prefix stripped, bare ids untouched, and dots preserved (LM Studio ids such as `qwen3.8-flash-next` carry dots).

## How to Test

1. Before the fix, the new tests fail:
   ```
   pytest tests/hermes_cli/test_model_normalize.py -q -k LMStudio
   3 failed, 2 passed
   ```
2. After the fix they pass, and the file's existing cases are unaffected:
   ```
   pytest tests/hermes_cli/test_model_normalize.py -q
   28 passed
   ```
3. End to end: set `model.default: lmstudio/<a-model-you-have-downloaded>` in `config.yaml` against a running LM Studio server. Before the fix every turn fails with the non-retryable 400 above; after it, the request carries the bare id and succeeds.

### Code

- [ ] I've run `pytest tests/ -q` and all tests pass — I ran `tests/hermes_cli/test_model_normalize.py` (28 passed); I did not run the full suite, which needs dev extras I don't have installed here. The change adds one string to a frozenset and touches no other module.
- [x] I've tested on my platform: macOS 15 (Darwin 25.5), Python 3.11, LM Studio native server

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A, the change is a set member with an inline comment
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — N/A, no new config keys
- [x] I've considered cross-platform impact — N/A, pure string normalization, no platform-specific paths

### DIFF
--- a/hermes_cli/model_normalize.py
+++ b/hermes_cli/model_normalize.py
@@ -104,6 +104,11 @@ _MATCHING_PREFIX_STRIP_PROVIDERS: frozenset[str] = frozenset({
     "xiaomi",
     "arcee",
     "ollama-cloud",
+    # LM Studio's native server validates the model id against the ids it has
+    # downloaded and rejects a provider-prefixed name outright with a
+    # non-retryable HTTP 400 (``Invalid model identifier "lmstudio/<id>" ...
+    # code=model_not_found``), so every turn dies until the prefix is stripped.
+    "lmstudio",
     "nebius-token-factory",
     "custom",
     "gemini",
--- a/tests/hermes_cli/test_model_normalize.py
+++ b/tests/hermes_cli/test_model_normalize.py
@@ -223,3 +223,30 @@ class TestLMStudioMatchingPrefixStrip:
             == "qwen3.8-flash-next"
         )
 
+    @pytest.mark.parametrize("model", [
+        "openai/gpt-4",
+        "z-ai/glm-5.2",
+        "ollama/qwen3",
+    ])
+    def test_foreign_prefix_untouched(self, model):
+        """Only the *matching* prefix is stripped.
+
+        An LM Studio endpoint can front a proxy whose route requires a
+        foreign prefix (LiteLLM-style ``ollama/…``), so a non-matching prefix
+        must survive — adding ``lmstudio`` to the strip set must not widen
+        into a general slash-stripper.
+        """
+        assert normalize_model_for_provider(model, "lmstudio") == model
+
+    @pytest.mark.parametrize("model,expected", [
+        ("lmstudio/lmstudio/qwen3-27b", "lmstudio/qwen3-27b"),
+        ("lmstudio/openai/gpt-4", "openai/gpt-4"),
+    ])
+    def test_exactly_one_prefix_level_is_stripped(self, model, expected):
+        """A stacked id loses only its own leading ``lmstudio/``.
+
+        Matches the one-level scope the other strip paths use: a leftover
+        prefix chain means the id was namespaced elsewhere, and guessing
+        further would mangle a legitimate route.
+        """
+        assert normalize_model_for_provider(model, "lmstudio") == expected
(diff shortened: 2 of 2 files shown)


Enough1122:

**98435 — Add `lmstudio` to model normalization stripping**

- `hermes_cli/model_normalize.py:9-13` adds `"lmstudio"` to the provider set that strips matching prefix. Rationale in comment: LM Studio validates model id and rejects `lmstudio/<id>` with 400 `model_not_found` — analogous to `ollama-cloud` already covered. Minimal, correct.
- `tests/hermes_cli/test_model_normalize.py:28-60` adds `TestLMStudioMatchingPrefixStrip` with parametrized cases for prefix strip, bare id untouched, and dotted id preserved (`qwen3.8` not mangled). Good coverage of the reported failure.

Non-blocking:
- No change to generic `normalize_model_for_provider` logic, so risk is confined to adding one entry; no new branching. **Non-blocking**.
- Consider adding a case for already-prefixed vs double-prefix, but out of scope. **Non-blocking**.

Verdict: Correct one-line fix with targeted regression tests.

---

Issue #81703, opened 2026-08-08
[Bug]: Desktop — Grok xAI OAuth final answer rendered twice after tool-heavy turn


networthexplained:
## Bug Description

On **Hermes Desktop**, with **Grok via xAI OAuth (SuperGrok subscription)**, the assistant’s **final answer after a tool-heavy turn is shown twice** in the transcript (two full copies of the same final prose, or a sealed interim copy + a second clean final).

This is **not** the model billing twice for a second generation in the observed case: the profile `state.db` holds **one** `role=assistant` row with `finish_reason=stop` for that turn. The duplicate is a **Desktop settle / dual-bubble** problem (same family as #63679 / #81422).

**Impact:** every long Grok research/tool turn becomes unreadable and looks like double delivery; user loses trust in subscription billing and in Desktop transcript fidelity.

## Steps to Reproduce

1. Open Hermes **Desktop** on macOS.
2. Start a new chat with model **`grok-4.5`**, provider **`xai-oauth`** (Grok / SuperGrok subscription path).
3. Send a prompt that forces a **long multi-tool turn** (many tool rounds, then one long final markdown answer). Example class: multi-page website review with skill loads + browser/fetch tools.
4. Wait until the turn fully completes (tools finish, final answer settles).
5. Observe the assistant final in the transcript.

## Expected Behavior

- Exactly **one** visible final assistant bubble for the turn’s answer.
- Interim tool-boundary commentary (if enabled) either stays clearly interim or is **settled in place** into that single final — never appended as a second full copy of the same prose.
- Backend and UI agree: one `finish_reason=stop` row → one UI final.

## Actual Behavior

- The final answer **appears twice** in Desktop after the tool-heavy Grok turn.
- Profile DB check on the same session: **a single** assistant message with `finish_reason=stop` and one copy of the final text (no end-to-end doubled content inside that row; no second stop row).
- So this is **UI dual-render / failed interim settle**, not double model completion and not double DB persist.

## Affected Component

- [x] Other → **Desktop** (`apps/desktop`) message stream settle path
- [x] Agent Core / Gateway (interim + `message.complete` / `response_previewed` signaling) — only as the event source Desktop mis-settles

**Surface:** Hermes Desktop chat transcript  
**Provider path:** `xai-oauth` · model `grok-4.5`  
**Not:** CLI-only (reporter surface is Desktop)

## Root Cause Analysis

[...]

Issue #81422, opened 2026-08-08
[Bug]: Desktop — assistant final answer rendered twice after tool-heavy turn (Grok / xAI OAuth)


networthexplained:
## Bug Description

On **macOS Desktop**, after a multi-step tool turn on **Grok 4.5 via xAI OAuth (subscription)**, the **same final assistant answer appears twice** in the transcript (two full copies back-to-back).

Backend session store for that turn has **one** final assistant row (`finish_reason=stop`, single copy of the text). So this is a **UI / stream-settle dual-bubble** failure, not the model writing the answer twice into history.

Related family: prior dual-render work around interim + complete (`#63679` — “partial first copy + clean final”). This report is the still-shipping case where a long final after many tool rounds still shows **two identical finals**.

## Steps to Reproduce

1. Open Hermes Desktop on macOS.
2. Use **Grok 4.5** with provider **xAI OAuth** (subscription path).
3. Send a research-style prompt that forces several tool rounds (skill load, multiple fetches, browser, then a long structured final).
4. Wait until the turn completes and the final prose answer is fully shown.
5. Scroll the transcript for that turn’s final answer.

## Expected Behavior

- Exactly **one** assistant bubble with the final answer for the turn.
- Cumulative transcript size grows roughly **linearly** with turns (not two copies of the same final per turn).
- Backend store and live UI stay in agreement (one final row → one visible final bubble).

## Actual Behavior

- The **same final answer is shown twice** in the Desktop UI for that turn.
- Session DB inspection after the fact: **one** assistant final with that text (not two identical finals in `messages`).
- Observed on a tool-heavy Desktop session using `grok-4.5` / `xai-oauth`.

## Affected Component

- [x] Other → **Desktop** (`apps/desktop` message stream / chat settle)
- [ ] Agent Core (secondary: only if dual `message.complete` / interim emit is proven; primary evidence points at UI settle)

## Root Cause Analysis (code runway)

Suspects align with the streaming-bubble dual-render class:

| Area | Path | Why it fits |
|------|------|-------------|
| Live bubble id reuse / soft start | `apps/desktop/src/app/session/hooks/use-message-stream/index.ts` — `mutateStream`: `const streamId = state.streamId ?? nextStreamMessageId('assistant-stream')` | If `streamId` is lost mid-turn after interim seals, a **new** stream bubble is seeded while an earlier sealed copy already holds the same text. |
[...]

Issue #71857, opened 2026-07-26
[Bug]: Desktop — identical assistant final answer delivered twice


networthexplained:
## Bug Description

The same assistant final answer is delivered **twice, identically** (same text, not a paraphrase). The user sees two copies of one completed reply in the conversation UI.

This is a message-delivery / transcript duplication bug, not “similar advice on a follow-up.”

## Steps to Reproduce

1. Open Hermes Desktop chat and use a normal multi-turn session.
2. Send a user message that produces a long assistant final answer.
3. Send a follow-up message in the same session.
4. Observe the assistant output for that turn (and/or the prior turn) in the transcript.

Frequency: observed in active Desktop use; treat as intermittent until narrowed.

## Expected Behavior

- Exactly **one** assistant final message bubble per completed agent turn.
- Transcript stores and renders a single copy of that final text.

## Actual Behavior

- The assistant final answer appears **twice with identical content**.
- Both copies are visible to the user as separate bubbles (or equivalent double presentation).
- Content is byte-for-byte the same from the user’s perspective, not a rephrased second answer.

## Affected Component

- [x] Desktop (conversation transcript / streaming / hydration)
- [x] Agent Core / session message persistence (if the duplicate is stored twice, not only painted twice)
- [ ] CLI (unknown — not confirmed)
- [ ] Gateway (unknown — not confirmed on this report)

## Related issues

Search hit several duplication reports; this filing is for **identical double final answers in Desktop chat**:

- #38319 — Duplicate messages in desktop app (hydration race with active stream)
- #37470 — Desktop: conversation history duplicated in responses
- #65293 — Previous conversation content duplicated after resume

Please dedupe against #38319 if maintainers confirm the same `streamId` / hydration race; keep open or cross-link if the duplicate is persisted in session history rather than only a UI double-paint.

## Root Cause Analysis (hypothesis)

Most likely Desktop stream + hydration race (as in #38319):

1. Live stream creates an assistant message keyed by `streamId`.
2. `hydrateFromStoredSession` (or equivalent) replaces `messages` without clearing an orphaned `streamId`.
3. Later `message.delta` / finalize cannot find `streamId` → inserts a **second** assistant message with the same final text.

Alternate: session layer appends the completed assistant message twice (persist path), so reload and live view both show twins.

Areas to inspect:

- `apps/desktop` — session hydration vs live stream (`use-message-stream` / desktop controller patterns cited in #38319)
- Session transcript write path — single append on `message.end` / completion
[...]

PR #98441, opened 2026-08-30
fix(tui): fold identical complete onto last interim

eloklam:
## What does this PR do?

This contribution was implemented, tested, and prepared by Hermes Agent on behalf of Enoch Lam (@eloklam). I am Hermes Agent.

Hermes TUI rendered the same assistant reply twice after a tool-heavy turn (especially Grok / xAI OAuth on the Responses path). The backend stored one `finish_reason=stop` row; the TUI still painted two bubbles because `message.complete` refused to settle onto a sealed interim unless `response_previewed` was set. Commentary turns correctly leave that flag false, so identical finals duplicated.

Desktop already folds when the last interim and the complete payload continue each other (exact or prefix-either-way). This PR brings the same settle to Ink TUI, and honours `already_streamed` so a truncated interim payload cannot erase tokens already painted by `message.delta`.

Distinct commentary that does not continue the final (for example pre-diff “Before edit.” and post-diff “After edit.”) still stays as its own bubble.

## Related Issue

Related Desktop dual-bubble reports: #81422, #81703, #71857. This PR does not claim those issues: they are Desktop settle paths. This change is the TUI counterpart (`already_streamed` + identical/prefix fold without waiting for `response_previewed`).

## Type of Change

- [ ] 🔒 Security fix
- [ ] 📝 Documentation update

## Changes Made

- `ui-tui/src/app/turnController.ts` — fold last sealed narration onto `message.complete` when texts continue; keep the longer copy; flush leftover deltas before fold; honour `already_streamed` on interim seal
- `ui-tui/src/app/createGatewayEventHandler.ts` — pass `already_streamed` into `recordInterimMessage`
- `ui-tui/src/__tests__/createGatewayEventHandler.test.ts` — identical and prefix settle without `response_previewed`; truncated `already_streamed` prefix keeps streamed tokens; stub Ink OSC 11 listeners so the suite can construct the handler in Vitest

## How to Test

1. `cd ui-tui && npx vitest run src/__tests__/createGatewayEventHandler.test.ts` (103 passed on this change)
2. `cd ui-tui && npm run typecheck`
3. In `hermes --tui`, run a tool-heavy Grok / xAI turn that previously showed the same final twice. Expect one assistant bubble. Mid-turn commentary that is not a prefix of the final should remain a separate segment.

### Code

- [ ] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [ ] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [ ] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
- [ ] My PR contains **only** changes related to this fix/feature (no unrelated commits)
- [ ] I've added tests for my changes (required for bug fixes, strongly encouraged for features)
- [ ] I've tested on my platform: 

### DIFF
--- a/ui-tui/src/__tests__/createGatewayEventHandler.test.ts
+++ b/ui-tui/src/__tests__/createGatewayEventHandler.test.ts
@@ -2084,7 +2084,7 @@ describe('createGatewayEventHandler', () => {
       expect(appended).toHaveLength(0)
     })
 
-    it('keeps identical interim and terminal replies as separate messages without response_previewed', () => {
+    it('settles identical terminal reply onto interim without response_previewed', () => {
       const appended: Msg[] = []
       const onEvent = createGatewayEventHandler(buildCtx(appended))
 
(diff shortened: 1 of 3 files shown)


Enough1122:

**98441 — TUI fold interim into terminal reply, preserve already_streamed tokens**

- `ui-tui/src/app/turnController.ts:89-129` adds helpers `lastAssistantNarrationIndex`, `textsContinue`, `foldContinuingInterim` to detect when last interim narration continues final text (exact or prefix either-way, trimmed). Folding picks longer of the two, preventing duplicate bubbles.
- `turnController.ts:142-168` seals `bufRef` via `flushStreamingSegment()` before fold, then folds last interim onto terminal reply. Dedupe logic updated: `dedupeStart` now 0 when `response_previewed` or folded, else `interimBoundaryIndex`. Sets `finalText=''` when folded to avoid re-append — matches comment "Do not append it again".
- `turnController.ts:176-199` extends `recordInterimMessage(text, alreadyStreamed)` to keep longer of `streamed` vs `authoritativeText` when `alreadyStreamed && textsContinue(...)`. Prevents truncated interim (`already_streamed:true, text:'Hello'` vs streamed `'Hello world'`) from erasing tokens already painted. Fallback to authoritative otherwise.
- `createGatewayEventHandler.ts:77` passes `Boolean(ev.payload?.already_streamed)` through.
- Tests `createGatewayEventHandler.test.ts:40-65` add three cases: identical settle to 1 msg, prefix-matched final settles, and truncated interim with `already_streamed` keeps full `"Hello world"`. Also mocks `@hermes/ink` foreground/background — unrelated but needed for isolation.

Non-blocking:
- `textsContinue` uses `trim()` both sides, so whitespace-only diffs treated as not continuing — reasonable. **Non-blocking**.
- Fold mutates `segmentMessages` slice copy, not original array — safe. **Non-blocking**.
- Edge: `finalTail` dedup after fold uses slice from `dedupeStart`; when folded, `finalText=''` skips tail — correct, but ensure `flushStreamingSegment` doesn't leave stale empty segment (guarded by `bufRef.trimStart()` check). **Non-blocking**.

Verdict: Fixes duplicate-message UX for Grok/xAI tool-heavy turns with careful test coverage.

---

Issue #90945, opened 2026-08-20
hermes update reports config current before new-version migration runs


miqcie:
## Bug description

After updating from Hermes Agent v0.19.1 to v0.20.4, `hermes update` printed `Configuration is up to date`, but an immediate `hermes doctor` reported `Config version outdated (v33 → v37)`. Running `hermes doctor --fix` then performed the migration.

This makes a successful update appear complete while leaving the configuration on the previous schema until the user discovers and runs a second command.

## Steps to reproduce

1. Start on v0.19.1 with config schema v33.
2. Run `hermes update` to v0.20.4.
3. Observe `Checking configuration for new options... Configuration is up to date`.
4. Run `hermes doctor`.
5. Observe `Config version outdated (v33 → v37)`.

## Expected behavior

The updater should run configuration migration using the newly installed code, or clearly tell the user that `hermes doctor --fix` is still required.

## Actual behavior

The updater claims configuration is current even though the new version considers it outdated.

## Environment

- macOS 27.0
- Upgrade: v0.19.1 → v0.20.4
- Config: v33 → v37

## Acceptance criteria

- An update that introduces a newer config schema migrates it before reporting completion, or reports the required follow-up accurately.
- A post-update `hermes doctor` no longer immediately contradicts the updater’s configuration status.

webtecnica:
## Diagnostic: already fixed on main (post-update config module reload)

Verified against current `main` (`298f662108`). The failure described — `hermes update` printing "Configuration is up to date" while the config schema is still on an old version — is exactly the pre-pull-process cache bug that `hermes_cli/update_cmd.py` now fixes:

1. **`_reload_config_modules()`** (update_cmd.py:88) force-reloads `hermes_cli.config_defaults`, `hermes_cli.config`, and `hermes_cli.config_migrations` from disk after `git pull`, so `DEFAULT_CONFIG["_config_version"]` reflects the freshly-pulled code instead of the old `sys.modules` cache. The docstring documents this exact failure: "check_config_version() reports (33, 33) — 'up to date' — even though the freshly-pulled code has v34 with a migration to run" (this was introduced for the #81946 personality-reset migration being silently skipped).

2. **`_run_config_check_fresh()`** (update_cmd.py:129) and **`_run_migrate_config_fresh()`** (update_cmd.py:141) both call `_reload_config_modules()` before reading the version / running the migration — the update path (update_cmd.py:5397 "CRITICAL" comment) never uses the cached module for version checks.

[...]

PR #98443, opened 2026-08-30
fix(wake): stop forcing onnxruntime/numpy downgrades for openWakeWord

### LINKED ISSUE #98440: [Bug] Windows: wake.openwakeword lazy install fails on DLL lock, leaves onnxruntime half-uninstalled and breaks local STT
## Summary
On Windows, enabling the wake word (`wake_word.enabled: true`) makes GUI
startup trigger a lazy install of `wake.openwakeword`. The spec pins
`onnxruntime==1.27.0`, but local STT (faster-whisper VAD) already uses
onnxruntime 1.28.0 in the same venv. The downgrade fails while the running
process holds the DLL (`拒绝访问 (os error 5)`), and the failed install
leaves onnxruntime half-uninstalled (`__init__.py` + 14 more files deleted),
which then breaks STT transcription with:
`module 'onnxruntime' has no attribute 'SessionOptions'`.

## Repro (Windows)
1. Install Hermes with local STT (faster-whisper → onnxruntime 1.28.0)
2. Set `wake_word.enabled: true` in config.yaml
3. Start desktop app / TUI (startup calls `wake.start(gui)` → `ensure('wake.openwakeword')`)
4. Lazy install fails: `failed to remove file ...onnxruntime_providers_shared.dll: 拒绝访问 (os error 5)`
5. Voice input now fails with the SessionOptions error (VAD loads onnxruntime)

## Root cause
`tools/lazy_deps.py` `wake.openwakeword` (mirrored in pyproject.toml `[wake]`
extra) pins `onnxruntime==1.27.0` even though openwakeword 0.6.0 declares
`onnxruntime <2,>=1.10.0`. Same downgrade-conflict class applies to
`numpy==2.4.3` (STT stack already installs 2.4.6). uv's failed uninstall
does not restore removed files, so a half-uninstalled package stays in
site-packages and breaks an unrelated feature.

Verified: openwakeword 0.6.0 loads and runs (VAD inference) on onnxruntime
1.28.0; with the spec relaxed, `ensure('wake.openwakeword')` performs no
install on a venv with onnxruntime 1.28.0 + numpy 2.4.6.

## Suggested fix
Relax the wake specs in `tools/lazy_deps.py` and pyproject.toml:
- `onnxruntime==1.27.0` → `onnxruntime>=1.10.0,<2`
- `numpy==2.4.3` → `numpy>=2.4.3,<3` (all three wake engines)

Optional hardening: after a failed lazy install, verify the affected
packages are still importable and restore from wheel if not — the
half-uninstall state is currently silent until an unrelated feature breaks.

## Environment
- Windows 11, Python 3.11.15 (uv venv)
- Hermes main @ 52279ea106

cmyyy:
## What does this PR do?

Fixes #98440. The `wake.openwakeword` lazy spec pinned `onnxruntime==1.27.0`
while local STT (faster-whisper VAD) uses onnxruntime 1.28.0 in the same
venv. On Windows, a runtime lazy install then tries to downgrade onnxruntime
while the running process holds the DLL, fails with
`拒绝访问 (os error 5)`, and leaves a half-uninstalled package behind —
breaking STT entirely (`module 'onnxruntime' has no attribute
'SessionOptions'`).

openwakeword 0.6.0 declares `onnxruntime <2,>=1.10.0`, so pinning 1.27.0 was
never required. Same downgrade-conflict class for `numpy==2.4.3` (the STT
stack already brings 2.4.6): relaxed to `>=2.4.3,<3` across all three wake
engines (openwakeword / sherpa / porcupine), so `ensure()` reuses installed
versions instead of forcing a downgrade.

## Test plan

- `ensure('wake.openwakeword')` performs no install on a venv with
  onnxruntime 1.28.0 + numpy 2.4.6 (previously it attempted a downgrade)
- openwakeword 0.6.0 loads and runs (VAD inference) on onnxruntime 1.28.0
- `tests/tools/test_lazy_deps*.py` + `tests/tools/test_wake_word.py`: 105
  passed, 4 platform skips; the one failure
  (`test_lazy_deps_durable_target.py::test_readonly_target_reports_error`)
  also fails on main — Windows read-only directory semantics

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -222,12 +222,15 @@ voice = [
 # first /wake; mirrored in tools/lazy_deps.py.
 wake = [
   "openwakeword==0.6.0",
-  "onnxruntime==1.27.0",
+  # Match openwakeword's own constraint (onnxruntime <2,>=1.10.0) instead of
+  # pinning 1.27.0, which forces a downgrade of the onnxruntime used by local
+  # STT and fails on Windows while the venv DLL is held by a running process.
+  "onnxruntime>=1.10.0,<2",
   "sherpa-onnx==1.13.4",
   "sentencepiece==0.2.2",
   "pvporcupine==4.0.3",
   "sounddevice==0.5.5",
-  "numpy==2.4.3",
+  "numpy>=2.4.3,<3",
   # openWakeWord's onnx embedding model scores near-zero on macOS ARM64
   # (dscripka/openWakeWord#336), so the wake word runs on tflite there.
   # Upstream declares tflite-runtime for Linux only; ai-edge-litert is the
--- a/tools/lazy_deps.py
+++ b/tools/lazy_deps.py
@@ -167,9 +167,14 @@ LAZY_DEPS: dict[str, tuple[str, ...]] = {
     ),
     "wake.openwakeword": (
         "openwakeword==0.6.0",
-        "onnxruntime==1.27.0",
+        # Match openwakeword's own constraint (onnxruntime <2,>=1.10.0). The
+        # previous ==1.27.0 pin forced a downgrade of the onnxruntime already
+        # installed for local STT (faster-whisper VAD), which fails on Windows
+        # while the running process holds the DLL, leaving a half-uninstalled
+        # package behind and breaking STT.
+        "onnxruntime>=1.10.0,<2",
         "sounddevice==0.5.5",
-        "numpy==2.4.3",
+        "numpy>=2.4.3,<3",
     ),
     # Open-vocabulary keyword spotting: any typed phrase, zero training.
     # sentencepiece is required by sherpa_onnx.text2token (runtime phrase
@@ -178,12 +183,12 @@ LAZY_DEPS: dict[str, tuple[str, ...]] = {
         "sherpa-onnx==1.13.4",
         "sentencepiece==0.2.2",
         "sounddevice==0.5.5",
-        "numpy==2.4.3",
+        "numpy>=2.4.3,<3",
     ),
     "wake.porcupine": (
         "pvporcupine==4.0.3",
         "sounddevice==0.5.5",
-        "numpy==2.4.3",
+        "numpy>=2.4.3,<3",
     ),
 
     # ─── Image generation backends ─────────────────────────────────────────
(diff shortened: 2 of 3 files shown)


Enough1122:

**98443 — Surface config-module reload failures & relax wake deps**

- `hermes_cli/update_cmd.py:10-36` changes `_reload_config_modules()` to return `list[str]` of failed modules, upgrades `logger.debug` to `logger.warning` for reload exceptions — surfaces stale-version/missed migration (#90945) instead of swallowing.
- `_run_config_check_fresh:50-66` prints warning `Could not reload updated config modules: ...` and hint `hermes doctor --fix` when `failed` non-empty, still returns `check_config_version()` result but warns it may be stale. Good UX.
- `_run_migrate_config_fresh:80-91` now refuses to migrate when reload failed, returns `{"skipped":True, "reason":"config-module reload failed"}` with printed hint — prevents applying wrong schema step against stale modules. Correct safety.
- Tests `tests/hermes_cli/test_update_config_reload_warning.py:53-86` cover: failed collection, empty on success, warning on check, silent on success, skip on migrate, success path. Uses `patch` on `importlib.reload` and `_reload_config_modules` — thorough.
- Dependency relax: `pyproject.toml:105-113` and `tools/lazy_deps.py:218-243` change `onnxruntime==1.27.0` → `>=1.10.0,<2` and `numpy==2.4.3` → `>=2.4.3,<3` for `wake` extras, matching `openwakeword` constraint `<2,>=1.10.0`. Comment notes previous pin forced downgrade of `onnxruntime` used by local STT (faster-whisper VAD) and failed on Windows while DLL held. Sensible.

Non-blocking:
- `uv.lock:265-292` churn includes `numpy` marker splits and `vercel-workers` dep markers — appears from `uv lock` regeneration, not intentional but harmless. Keep an eye on unrelated lock diffs. **Non-blocking**.
- Printed warnings use `print()` not logger; acceptable for CLI `hermes update` context but consider routing through `rich` console for consistency. **Non-blocking**.

Verdict: Important reliability fix for post-update config migration with well-argued dep relaxation.

---

cmyyy:
Quick note on the blocked state: the PR originally carried an unrelated local commit (`52279ea`, the config-module reload fix for #90945, which is not on upstream main) because the branch was cut from a stale local checkout. I have rebased onto the latest upstream main and force-pushed — the PR now contains a single commit (`5cd0c67e`) touching 3 files (`tools/lazy_deps.py`, `pyproject.toml`, `uv.lock`).

Heads-up: the automated AI review above reviewed the old 5-file diff, including that unrelated commit; the current diff is only the wake dependency relaxation. Ready for CI and review.

PR #98445, opened 2026-08-30
fix(gateway): mark shutdown/restart notifications as interim sends

### LINKED ISSUE #98432: [Bug]: shutdown/restart notification is sent without the interim marker its own contract requires
## Bug Description

`_interim_metadata` (`gateway/run.py`) states a contract:

> Mark a mid-turn status/advisory send as NOT the turn-final.
>
> Stream-is-the-message adapters (relay Slack native streaming) intercept the first
> unmarked send to an armed (chat, turn) key and seal the live stream with its content.
> **Every gateway-side send that can fire while a turn is streaming — heartbeats,
> inactivity warnings, approval fallbacks, background-review notices — MUST carry this
> marker** or it will seal the user's answer stream with status text

`_notify_active_sessions_of_shutdown()` is, by construction, a send that fires while turns
are running: its recipient list is `_snapshot_running_agents()`. It does not carry the
marker.

Its three `adapter.send()` sites pass either the thread-routing metadata alone, or nothing
at all:

```python
msg = f"⚠️ Gateway {action} — {hint}"
...
result = await adapter.send(chat_id, msg, metadata=metadata)              # thread routing only
result = await adapter.send(str(home.chat_id), msg, metadata=metadata)    # thread routing only
result = await adapter.send(str(home.chat_id), msg)                       # no metadata at all
```

By the docstring's own account, on a stream-is-the-message adapter this seals the user's
in-flight answer with the shutdown text.

## Steps to Reproduce

Deterministic and offline, inside the released image
(`nousresearch/hermes-agent:v2026.8.27`). No gateway, no session, no network:

```python
import inspect, re, sys
sys.path.insert(0, "/opt/hermes")
import gateway.run as r

src = inspect.getsource(r.GatewayRunner._notify_active_sessions_of_shutdown)
print("carries _interim_metadata :", "_interim_metadata" in src)
print("adapter.send call sites   :", len(re.findall(r"adapter\.send\(", src)))
for line in src.splitlines():
    if "adapter.send(" in line or "msg = f" in line:
        print("   ", line.strip())

whole = inspect.getsource(r)
print("_interim_metadata used elsewhere in the module:", whole.count("_interim_metadata(") - 1)
```

Output:

```
carries _interim_metadata : False
adapter.send call sites   : 4
    msg = f"⚠️ Gateway {action} — {hint}"
    result = await adapter.send(chat_id, msg, metadata=metadata)
    result = await adapter.send(str(home.chat_id), msg, metadata=metadata)
    result = await adapter.send(str(home.chat_id), msg)
_interim_metadata used elsewhere in the module: 4
```

Three send sites, none marked, in a module that uses the marker four times elsewhere.

## Second-order effect: the message cannot currently be localized

Without the marker, an adapter that must treat this message differently from a normal
reply has only one thing to key on: **the English text**.

That is not hypothetical. An out-of-tree platform adapter has to distinguish this notice
from a turn-final reply, because closing the turn as `complete` on a shutdown notice would
record a truncated turn as a clean one. With no metadata to inspect, it matches on the
literal prefixes `"⚠️ Gateway shutting down"` and `"⚠️ Gateway restarting"` — a copy of your
string living outside your repo, which silently stops matching the day the wording changes.

Two things are already moving against that:

- **[PR #57037](https://github.com/NousResearch/hermes-agent/pull/57037)** (open) appends a
  reason clause to this very message;
- the string is hardcoded while its immediate neighbour in the same sequence is not. A
  `/restart` on a non-English gateway emits `t("gateway.restart.restarting")` — translated,
  46 `gateway.*` keys exist in `it.yaml` alone — immediately followed by this f-string, in
  English. The last line the user reads is the untranslated one.

So localizing it today would break every adapter that recognizes it by text. The marker is
what makes the translation safe, which is why these are one report and not two.

## Expected Behavior

That a shutdown/restart notice be distinguishable from a turn-final reply by something
other than its wording — as the module's own contract requires for sends of this class.
Which mechanism is the maintainers' call.

## Version

  release, and where the reproduction above was run.
- Still present on `main` at `2635035` (2026-08-30 03:57 UTC): the message at
  `gateway/run.py:11137-11144`, the send sites at `:11219` and `:11302`.

```python
src = inspect.getsource(r.GatewayRunner._notify_active_sessions_of_shutdown)
assert "_interim_metadata" in src
```

And on the wire: a shutdown notice arriving at an adapter with no field that separates it
from the turn's own reply.

## Not covered by existing reports

Four issues and one PR touch this message — the shutdown reason (#57037), re-broadcasting
(#71180), recovery routing (#89396), listing interrupted tasks (#74755) — and none of them
concerns the marker or the wording's translatability. [#70732](https://github.com/NousResearch/hermes-agent/issues/70732)
covers hardcoded English strings as a class and names `gateway/run.py`, but not this send
and not the marker, which is the part that has to come first.

liuhao1024:
## What does this PR do?

`_notify_active_sessions_of_shutdown` fires at the very start of `stop()`, while turns may still be streaming — its recipient list is `_snapshot_running_agents()`. Per the contract in `_interim_metadata`'s docstring, every gateway-side send that can fire mid-turn MUST carry the interim marker, or a stream-is-the-message adapter (relay Slack native streaming) will intercept the first unmarked send to an armed `(chat, turn)` key and seal the user's live answer stream with the status text.

The three `adapter.send` sites in this function passed only thread-routing metadata, or no metadata at all — exactly the contract violation #98432 describes. This PR routes all three sends through `_interim_metadata` so the shutdown/restart advisory can never be mistaken for the turn-final. The home-channel branch collapses its `metadata`/no-`metadata` if/else into a single interim-marked call (`_interim_metadata(None)` yields `{"_interim_send": True}`).

The marker is gateway-internal and already stripped before the wire by the relay adapters (`gateway/relay/adapter.py` pops `_interim_send` in both egress doors), and four other call sites in `gateway/run.py` already use `_interim_metadata` the same way — this brings the shutdown path in line with the established pattern.

## Related Issue

Fixes #98432

## Type of Change

## Changes Made

- `gateway/run.py` — `_notify_active_sessions_of_shutdown`: wrap the per-active-session send's metadata in `_interim_metadata(...)`; collapse the home-channel if/else into one send that always carries `_interim_metadata(metadata)`
- `tests/gateway/test_restart_notification.py` — updated the existing thread-metadata assertion to expect `_interim_send: True`; added `test_shutdown_home_channel_broadcast_carries_interim_marker` covering both the per-session send and the no-thread-metadata home-channel broadcast

## How to Test

1. `python -m pytest tests/gateway/test_restart_notification.py -q` — Observed result: 13 passed
2. `python -m pytest tests/gateway/test_gateway_shutdown.py tests/gateway/test_interim_send_lanes.py tests/gateway/test_cron_interrupt_notification.py tests/gateway/test_stream_final_contract.py tests/gateway/test_shutdown_cache_cleanup.py -q` — Observed result: 34 passed
3. Negative anchor: with the `gateway/run.py` change stashed (tests only), both the updated assertion and the new test fail — confirming they pin the interim marker, not incidental behavior. Observed result: 2 failed, 11 passed
4. `python -m ruff check gateway/run.py tests/gateway/test_restart_notification.py` — All checks passed

### Code

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -11216,7 +11216,12 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                     adapter=adapter,
                 )
 
-                result = await adapter.send(chat_id, msg, metadata=metadata)
+                # Fires while a turn may still be streaming — must carry the
+                # interim marker so a stream-is-the-message adapter doesn't
+                # seal the user's in-flight answer with this advisory (#98432).
+                result = await adapter.send(
+                    chat_id, msg, metadata=_interim_metadata(metadata)
+                )
                 if result is not None and getattr(result, "success", True) is False:
                     logger.debug(
                         "Failed to send shutdown notification to %s:%s: %s",
@@ -11296,10 +11301,12 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                     home.thread_id,
                     adapter=adapter,
                 )
-                if metadata:
-                    result = await adapter.send(str(home.chat_id), msg, metadata=metadata)
-                else:
-                    result = await adapter.send(str(home.chat_id), msg)
+                # Same interim-marker requirement as the per-session sends
+                # above: the broadcast races live turns on stream-is-the-
+                # message adapters (#98432).
+                result = await adapter.send(
+                    str(home.chat_id), msg, metadata=_interim_metadata(metadata)
+                )
                 if result is not None and getattr(result, "success", True) is False:
                     logger.debug(
                         "Failed to send shutdown notification to home channel %s:%s: %s",
--- a/tests/gateway/test_restart_notification.py
+++ b/tests/gateway/test_restart_notification.py
@@ -388,8 +388,45 @@ async def test_shutdown_notifications_use_cached_live_thread_source_when_origin_
     adapter.send.assert_awaited_once_with(
         "parent-42",
         "⚠️ Gateway shutting down — Your current task will be interrupted.",
-        metadata={"thread_id": "topic-7"},
+        metadata={"thread_id": "topic-7", "_interim_send": True},
     )
+
+
+@pytest.mark.asyncio
+async def test_shutdown_home_channel_broadcast_carries_interim_marker():
+    """The home-channel shutdown broadcast must be an interim send (#98432).
+
+    ``_notify_active_sessions_of_shutdown`` fires at the start of stop(),
+    while turns may still be streaming. Stream-is-the-message adapters
+    (relay Slack native streaming) intercept the first *unmarked* send to an
+    armed (chat, turn) key and seal the live stream with its content — so the
+    shutdown advisory would otherwise seal the user's in-flight answer with
+    status text. Both the per-active-session sends and the home-channel
+    broadcast must carry the ``_interim_send`` marker, including when no
+    thread routing metadata exists.
+    """
+    runner, adapter = make_restart_runner()
+    source = make_restart_source(chat_id="active-42", chat_type="dm")
+    session_key = build_session_key(source)
+
+    runner.config.platforms[Platform.TELEGRAM].home_channel = HomeChannel(
+        platform=Platform.TELEGRAM,
+        chat_id="home-42",
+        name="Ops Home",
+    )
+    runner._running_agents[session_key] = object()
+    runner.session_store._entries[session_key] = MagicMock(origin=source)
+    adapter.send = AsyncMock(return_value=SendResult(success=True, message_id="m"))
+
+    await runner._notify_active_sessions_of_shutdown()
+
+    targets = [c.args[0] for c in adapter.send.await_args_list]
+    assert targets == ["active-42", "home-42"]
+    for call in adapter.send.await_args_list:
+        assert call.kwargs["metadata"]["_interim_send"] is True, (
+            f"send to {call.args[0]} must be marked interim, got "
+            f"metadata={call.kwargs['metadata']}"
+        )
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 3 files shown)


Enough1122:

**98445 — Mark shutdown notifications as interim sends**

- `gateway/run.py:12-15,27-32` wraps both per-session and home-channel shutdown sends with `_interim_metadata(metadata)` so `metadata` carries `"_interim_send":True`. Comment ties to race: `_notify_active_sessions_of_shutdown` fires while turns may still stream; stream-is-the-message adapters (relay Slack native streaming) seal live stream on first unmarked send — without marker the advisory would clobber in-flight answer (#98432). Correct.
- Tests: `test_restart_notification.py:42-83` updates existing assertion to expect `{"thread_id":"topic-7","_interim_send":True}` and adds `test_shutdown_home_channel_broadcast_carries_interim_marker` asserting both `active-42` and `home-42` targets carry `"_interim_send":True` even when no thread metadata. Also `test_restart_resume_pending.py:97-100` updates `sent_calls` expectations to include `"_interim_send":True` (second call previously `None` now `{"_interim_send":True}`).

Non-blocking:
- `_interim_metadata` helper not shown in diff context but used symmetrically; ensure it handles `None` → `{"_interim_send":True}` (test proves it does). **Non-blocking**.
- No change to shutdown flow ordering — marker is additive, low risk. **Non-blocking**.

Verdict: Small, well-tested fix for a subtle streaming race.

---

liuhao1024:
Thanks for the review! Yes — `_interim_metadata(None)` returns `{"_interim_send": True}`, which is exactly why the home-channel `if metadata/else` branches collapse into a single `send(..., metadata=_interim_metadata(metadata))` call: no thread metadata is needed for the marker to apply. Marker is purely additive to the shutdown flow ordering.

PR #98446, opened 2026-08-30
fix(desktop): add "load more" to Command Center sessions list

### LINKED ISSUE #98444: Command Center Sessions list has no "load more" — sessions beyond the first 50 are unreachable from it
# Issue: Command Center Sessions list has no "load more" — sessions beyond the first 50 are unreachable from it

## Bug Description

The Command Center **Sessions** tab only ever shows the sessions that are
already loaded into the shared `$sessions` store. That store is paginated by
the sidebar with a default window of `SIDEBAR_SESSIONS_PAGE_SIZE` = 50 rows.
The sidebar itself exposes a "load more" control (`SidebarLoadMoreRow` →
`bumpSessionsLimit()` + refetch) that grows the window, but the Command
Center has **no equivalent affordance** — so once you have more than ~50
active sessions, the older ones are unreachable from the Command Center
(they are not listed, and there is no way to page deeper).

## Steps to Reproduce

1. Hermes Desktop with more than 50 active (non-archived) sessions.
2. Open the Command Center (status bar / Cmd+K) → **Sessions** tab.
3. Scroll the session list to the bottom.
4. Observed: only the first ~50 (the sidebar's loaded window) are shown, and
   there is **no "load more" button** at the bottom — the list just ends.
5. Compare with the app's left sidebar, which *does* show a load-more control
   for the same recents list.

## Expected Behavior

The Command Center Sessions list should offer the same "load more" affordance
as the sidebar (clicking it grows the shared `$sessions` window by 50 and
refetches), so all sessions are reachable from it too.

## Actual Behavior

No load-more button anywhere in the Command Center Sessions list; the list is
hard-capped at whatever the sidebar has already loaded (default 50).

## Environment

- Surface: Hermes Desktop (Command Center → Sessions)
- OS: Windows 10 (likely all desktop platforms — this is renderer-only logic)

## Notes / Related

- Root cause tracked in the code: `apps/desktop/src/app/command-center/index.tsx`
  renders `$sessions` directly with no pagination, while the sidebar grows the
  same store via `bumpSessionsLimit()`.
- Related existing issue: #51694 — "Command Center (⌘K) only searches loaded
  sidebar sessions — no FTS, no archived, no cross-profile", which calls out
  the same hard cap ("with 246 active sessions, only ~50 are searchable").
  This issue is narrower: the *browse list* itself lacks a load-more control,
  independent of search.

zqy1-1:
fix(desktop): add "load more" to Command Center Sessions list

## Changes

- `apps/desktop/src/app/command-center/index.tsx` (+12/-5):
  - Add `onLoadMoreSessions` prop (optional) and `loadMorePending` state
  - Gate `$sessionProfilesTruncated` subscription behind the Sessions tab
    with `useStoreSelector` + stable empty object, matching the existing
    re-render discipline for `$sessions` / `$pinnedSessionIds`
  - Show the real page size in `SidebarLoadMoreRow` so the label reads
    "load 50 more", consistent with the sidebar
  - Hide the load-more affordance when `onLoadMoreSessions` is not wired
  - Move button outside the scroll container (flex layout), fixed at the
    bottom-right extreme edge
- `apps/desktop/src/app/contrib/wiring.tsx` (+1): wire `loadMoreSessions`

## Why

The Command Center Sessions page reads the shared `$sessions` store
(paginated by the sidebar, default 50 items). The sidebar has a "load
more" button (`SidebarLoadMoreRow` + `bumpSessionsLimit()`), but the
Command Center had no equivalent — sessions beyond the first 50 were
unreachable from it. This adds the same mechanism.

Fixes #98444

### DIFF
--- a/apps/desktop/src/app/command-center/index.tsx
+++ b/apps/desktop/src/app/command-center/index.tsx
@@ -1,4 +1,3 @@
-import { useStore } from '@nanostores/react'
 import { type MouseEvent, type ReactNode, useCallback, useEffect, useMemo, useRef, useState } from 'react'
 
 import { LogTail } from '@/components/chat/log-tail'
@@ -30,7 +29,7 @@ import { fmtDateTime } from '@/lib/time'
 import { useStoreSelector } from '@/lib/use-session-slice'
 import { cn } from '@/lib/utils'
 import { upsertDesktopActionTask } from '@/store/activity'
-import { $pinnedSessionIds, pinSession, unpinSession } from '@/store/layout'
+import { $pinnedSessionIds, pinSession, SIDEBAR_SESSIONS_PAGE_SIZE, unpinSession } from '@/store/layout'
 import { $sessionProfilesTruncated, $sessions, sessionPinId } from '@/store/session'
 
 import { SidebarLoadMoreRow } from '../chat/sidebar/load-more-row'
@@ -56,6 +55,7 @@ type UsagePeriod = (typeof USAGE_PERIODS)[number]
 // component never re-renders from $sessions ticks while on System/Usage/etc.
 const EMPTY_SESSIONS: readonly never[] = []
 const EMPTY_PINNED: readonly string[] = []
+const EMPTY_TRUNCATED: Record<string, boolean> = {}
 
 interface CommandCenterViewProps {
   initialSection?: CommandCenterSection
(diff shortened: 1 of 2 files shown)


kokhlo:
Read the surrounding file while tracing this for #98444 — the fix direction is right, two gaps worth addressing before merge:

**1. Unconditional store subscription breaks the file's re-render discipline.** Lines 51-53 and the comment above the `sessions` selector ("we only need the data on the Sessions tab. Subscribe conditionally so the System/Usage/Maintenance tabs don't re-render on every stream delta") document why `$sessions` and `$pinnedSessionIds` go through `useStoreSelector` with a `section === 'sessions'` gate. The new `useStore($sessionProfilesTruncated)` subscribes unconditionally, so every truncated-map update now re-renders the Command Center on every tab. Same gate fixes it:

```tsx
const EMPTY_TRUNCATED: Record<string, boolean> = {}
const sessionProfilesTruncated = useStoreSelector(
  $sessionProfilesTruncated,
  s => (section === 'sessions' ? s : EMPTY_TRUNCATED)
)
```
(The stable-empty-array trick the file already uses for `EMPTY_SESSIONS` / `EMPTY_PINNED`.)

**2. `step={0}` drops the helpful label.** `SidebarLoadMoreRow` renders `t.sidebar.loadCount(step)` when `step > 0` — passing the real page size keeps the "load 50 more" affordance consistent with the sidebar: `step={SIDEBAR_SESSIONS_PAGE_SIZE}`.

Also a note on `onLoadMoreSessions` wiring: the prop is optional, but when it is missing the button renders with a click that no-ops (`if (!onLoadMoreSessions ...) return`). Gating the render on `hasMoreSessions && onLoadMoreSessions` would hide the dead affordance at call sites that don't wire it.

zqy1-1:
Thanks for the thorough review! All three points are valid — I've addressed them in a follow-up commit:

1. **Re-render discipline**: Wired `$sessionProfilesTruncated` through `useStoreSelector` with a stable empty object, matching the existing pattern for `$sessions` / `$pinnedSessionIds`. System/Usage/Maintenance tabs now won't re-render on session fetch ticks.

2. **Step value**: Changed `step={0}` to `step={SIDEBAR_SESSIONS_PAGE_SIZE}` so the label reads "load 50 more", consistent with the sidebar.

3. **Dead affordance**: The button is now gated on `!!onLoadMoreSessions` — no button when the prop isn't wired.

Diff: 12 insertions(+), 5 deletions(-) in `command-center/index.tsx`.

PR #98447, opened 2026-08-30
fix(browser): real-profile reuse probe no longer resurrects a browser

### LINKED ISSUE #98437: [Bug]: real-profile reuse probe can resurrect Chrome for Testing before signed Chrome
## Bug description

On current upstream `main`, the macOS real-profile reuse check still invokes an agent-browser command that can launch Chrome for Testing before Hermes launches the installed signed Google Chrome binary.

PR #98249 correctly changed the intended launch path to start the user's native browser on the Hermes-managed snapshot and attach agent-browser over CDP. However, the reuse check in `_real_profile_cdp()` first calls `_agent_browser_get_cdp(_REAL_PROFILE_SESSION)`. That helper runs:

```text
agent-browser --session hermes-real-profile get cdp-url
```

When the named session is not running, `get cdp-url` is not observational: agent-browser launches a browser and returns that new endpoint. A clean isolated session deterministically demonstrates that side effect. In the affected Hermes run, stale persisted session state was also present from an earlier real-profile generation; the probe resurrected Chrome for Testing on the managed copy with `--use-mock-keychain` and `--password-store=basic`. The copied macOS Keychain-backed Google cookies then became unusable, and Google Drive opened at the sign-in page.

This is a post-#98249 edge case in reuse discovery, not the original profile-selection or snapshot-copy failure.

## Minimal deterministic reproduction of the launching probe

The underlying behavior can be reproduced independently with a new isolated session that has no running browser:

```bash
SESSION="probe-absent-0351-$$"
npx -y agent-browser@0.35.1 --session "$SESSION" get cdp-url
npx -y agent-browser@0.35.1 --session "$SESSION" close
```

Observed output:

```text
[agent-browser] launched browser
ws://127.0.0.1:50751/devtools/browser/<redacted-id>
```

The probe session was closed afterward and no probe process remained.

## Observed end-to-end failure condition

The signed-out result was observed under this additional condition, which is important to distinguish from the clean isolated probe above:

1. Use macOS with signed stable Google Chrome as the default browser and an authenticated Chrome profile.
2. Use current Hermes upstream `main` at `26350357d76e4508c8df9304a3374bdc5a6f6220`.
3. Retain stale persisted `hermes-real-profile` agent-browser session state from an earlier real-profile generation, bound to the Hermes-managed copy, while no browser for that session is currently running.
4. Set `browser.use_real_profile=true`, `browser.engine=auto`, and remove any `browser.cdp_url` override.
5. Invoke the stock `browser_exec` real-profile path.
6. Observe the pre-snapshot reuse probe launch Chrome for Testing on the managed copy with `--use-mock-keychain` and `--password-store=basic`.
7. Navigate read-only to an authenticated Google service such as Google Drive; it opens signed out.

Without the stale managed-copy binding, the clean isolated probe still launches an unwanted throwaway browser, but the later `_cdp_on_data_dir()` check rejects it and `_agent_browser_close_session()` closes it before Hermes proceeds. Therefore the isolated probe proves the mutating discovery behavior; the stale persisted binding is what connected that behavior to the observed managed-copy authentication failure.

## Expected behavior

Reuse discovery must only observe an already-running real-profile browser. If the named session is absent, it should return no endpoint without launching anything. Hermes should then snapshot the selected source profile and launch the installed signed browser on that managed copy as intended by #98249.

## Actual behavior

- `_real_profile_cdp()` calls `_agent_browser_get_cdp()` before the signed-browser launch.
- `agent-browser get cdp-url` launches a browser when the named session is not running.
- For a clean wrong-directory launch, Hermes rejects and closes the session only after the launch has occurred.
- With stale persisted session state bound to the managed copy, the probe can resurrect Chrome for Testing using `--use-mock-keychain` and `--password-store=basic`; that was the condition under which Google Drive opened signed out.

The selected source profile was independently verified before the run:

- Chrome `Local State` selected `Profile 1`.
- The selected Cookies database passed SQLite integrity checks.
- It contained 558 cookie rows, including Google-domain authentication-cookie metadata.
- The prelaunch managed snapshot contained the expected cookie rows and non-empty encrypted values.

No cookie values, account identifiers, or credentials were inspected or included here. These checks rule out an empty, corrupt, or obviously wrong source profile. Authentication failed after the correct profile data was copied.

## Why the current attachment check does not prevent this

The current flow later checks whether the returned CDP port matches the managed copy's `DevToolsActivePort`. If the endpoint is on a wrong directory, it calls `_agent_browser_close_session()` before snapshotting. That is useful cleanup, but it occurs only after the supposedly observational `get cdp-url` command has already launched a browser and changed process/session state. It also does not reject a Chrome for Testing process when stale session state makes that process use the expected managed copy and port file.

Current call sequence in `tools/browser_tool.py`:

```text
_real_profile_cdp()
  -> _agent_browser_get_cdp(_REAL_PROFILE_SESSION)
       -> agent-browser --session ... get cdp-url
  -> _cdp_http_ready(existing)
  -> _cdp_on_data_dir(existing, copy_dir)
  -> _agent_browser_close_session(...) if the directory check fails
```

## Suggested direction

Make reuse discovery non-launching:

1. Observe the managed copy's `DevToolsActivePort` directly, or add/use an agent-browser status API that guarantees it never launches an absent session.
2. Before reuse, verify the endpoint belongs to the expected signed stable browser process using the exact managed `--user-data-dir`.
3. Reject Chrome for Testing and processes using `--use-mock-keychain` or `--password-store=basic`.
4. Fail closed if any unverified Chromium process owns the managed copy; do not overlay Cookies/Login Data while a browser may be using them.
5. Add a regression asserting that no agent-browser launch query occurs before the signed browser is started.

## Environment

- macOS 26.6.1, Apple Silicon
- Google Chrome 152.0.7977.64
- Hermes upstream `main`: `26350357d76e4508c8df9304a3374bdc5a6f6220`
- agent-browser installed versions checked: 0.34.0 and 0.35.1
- The independent absent-session probe above used agent-browser 0.35.1

## Safety and cleanup

- Testing was read-only; no forms, messages, uploads, settings, or external records were changed.
- The source Chrome profile was never launched under automation or modified.
- The managed browser and probe session were closed.
- `browser.use_real_profile` was returned to `false` after failed acceptance.
- The Hermes-managed copied profile was removed after confirming no process held it.
- The source Cookies database remained healthy after rollback.

Related: #96729, #98249.

chelsealong:
## What does this PR do?

Fixes the post-#98249 edge case in real-profile reuse discovery described in
#98437: `_real_profile_cdp()` probed for an already-running copy-browser by
calling agent-browser's `get cdp-url`. That command is **not observational**
— when the named session (`hermes-real-profile`) isn't already running,
agent-browser launches its own throwaway browser and returns that new
endpoint. With stale persisted session state from an earlier real-profile
generation, this probe resurrected Chrome for Testing on the Hermes-managed
profile copy with `--use-mock-keychain --password-store=basic` *before* the
intended signed-browser launch path ever ran. That makes every
keychain-encrypted cookie in the copy undecryptable, so the copy launches
signed out (Google Drive opening at the sign-in page, as reported).

The reuse probe now reads the copy dir's own `DevToolsActivePort` file
directly instead of asking agent-browser. Chrome writes that file itself
with its live debug port, and only while it's actually running on that exact
directory — so reading it is a pure observation with no side effects. If the
file is missing/stale (no readable HTTP endpoint on it), that now correctly
means "no live browser to reuse," not "launch one to find out." The
subsequent `_agent_browser_close_session()` best-effort cleanup call — itself
a `close`, never a launch — is now unconditional right before the
overlay+relaunch, so any stale agent-browser session-name state is cleared
either way.

`_agent_browser_get_cdp()` remains in use for the one place where a launch
*is* the correct behavior: attaching to the browser Hermes just launched
itself, after the signed-browser process already exists.

## Related Issue

Fixes #98437

## Type of Change

## Changes Made

- `tools/browser_tool.py`: added `_cdp_from_data_dir()`, a non-launching
  read of a copy dir's `DevToolsActivePort` file; `_real_profile_cdp()`'s
  reuse probe now calls it instead of `_agent_browser_get_cdp()`, and the
  stale-session `close()` call is now unconditional on a probe miss.
- `tests/tools/test_browser_real_profile.py`: added
  `test_reuse_probe_never_calls_agent_browser_get_cdp` (asserts the
  launching `get cdp-url` call happens at most once — only for the
  post-launch attach, never for reuse discovery) and
  `test_stale_port_file_is_closed_not_reused` /
  `test_cdp_from_data_dir_reads_devtoolsactiveport` for the new helper;
  updated the three existing tests that previously mocked the reuse probe
  through `_agent_browser_get_cdp`/`_cdp_on_data_dir` to reflect the new
  direct-file-read path.

## How to Test

1. `git checkout HEAD~1 -- tools/browser_tool.py` (revert only the source fix,
   keep the updated tests), then run:
   ```
   scripts/run_tests.sh tests/tools/test_browser_real_profile.py -k \
     "test_reuse_probe_never_calls_agent_browser_get_cdp or test_cdp_from_data_dir_reads_devtoolsactiveport"
   ```
   Both fail on the pre-fix code: the reuse probe calls
   `_agent_browser_get_cdp` twice (`assert get_cdp.call_count == 1` →
   `AssertionError: assert 2 == 1`), and `_cdp_from_data_dir` doesn't exist
   yet (`AttributeError`).
2. `git checkout HEAD -- tools/browser_tool.py` to restore the fix, then:
   ```
   scripts/run_tests.sh tests/tools/test_browser_real_profile.py
   ```
   → `79 tests passed, 0 failed`.
3. `ruff check tools/browser_tool.py tests/tools/test_browser_real_profile.py`
   → `All checks passed!`

### Code

- [x] I've run the test suite for the affected file and all tests pass
- [ ] I've tested on my platform — this fix was verified via unit tests only
      (macOS-specific reproduction from the issue was not re-run; the bug is
      platform-independent — agent-browser's `get cdp-url` launching behavior
      is not OS-specific — and is fully exercised by the added regression
      test)

- [x] N/A — no docs, config keys, or tool schemas changed

---

*This fix was prepared with AI assistance (Claude) and verified by running
the affected test suite locally before submission.*

### DIFF
--- a/tests/tools/test_browser_real_profile.py
+++ b/tests/tools/test_browser_real_profile.py
@@ -239,7 +239,7 @@ class TestRealProfileCdpLaunch:
              patch("hermes_cli.browser_connect.chromium_executable", return_value="/usr/bin/chrome"), \
              patch.object(bt.subprocess, "Popen", side_effect=fake_popen), \
              patch.object(bt, "_agent_browser_get_cdp",
-                          side_effect=[None, "http://127.0.0.1:41000"]), \
+                          return_value="http://127.0.0.1:41000"), \
              patch.object(bt, "_find_agent_browser", return_value="/usr/bin/agent-browser"), \
              patch.object(bt.subprocess, "run", return_value=proc), \
              patch.object(bt, "_is_headed_mode", return_value=False):
@@ -289,7 +289,7 @@ class TestRealProfileCdpLaunch:
              patch("hermes_cli.browser_connect.chromium_executable", return_value="/usr/bin/chrome"), \
              patch.object(bt.subprocess, "Popen", side_effect=fake_popen), \
              patch.object(bt, "_agent_browser_get_cdp",
-                          side_effect=[None, "http://127.0.0.1:41000"]), \
+                          return_value="http://127.0.0.1:41000"), \
              patch.object(bt, "_find_agent_browser", return_value="/usr/bin/agent-browser"), \
              patch.object(bt.subprocess, "run", side_effect=fake_run), \
              patch.object(bt, "_is_headed_mode", return_value=False):
@@ -302,12 +302,15 @@ class TestRealProfileCdpLaunch:
         assert "--cdp" in captured["argv"]
         self._reset()
 
-    def test_reuses_only_session_on_our_copy_dir(self, tmp_path):
-        """A live session on a DIFFERENT dir (stale/throwaway) is closed, not reused."""
+    def test_reuse_probe_never_calls_agent_browser_get_cdp(self, tmp_path):
+        """Reuse discovery must not invoke agent-browser's launching ``get
+        cdp-url`` before the signed browser has started (#98437): that
+        command launches a throwaway browser when the named session isn't
+        already running. The probe must resolve reuse purely from the copy
+        dir's own DevToolsActivePort file."""
         import tools.browser_tool as bt
         self._reset()
         proc = Mock(return_value=None, returncode=0, stdout="", stderr="")
-        closed = {"n": 0}
 
         class FakeChrome:
             def poll(self):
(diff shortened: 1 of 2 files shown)


Enough1122:

**98447 — Real-profile CDP reuse probe without launching throwaway browser**

- `tools/browser_tool.py:140-158` adds `_cdp_from_data_dir(data_dir)` reading `DevToolsActivePort` directly instead of calling `agent-browser get cdp-url` which launches a throwaway browser when named session not running (#98437). Returns `http://127.0.0.1:<port>` or `None` on missing/non-digit.
- `tools/browser_tool.py:171-185` rewrites `_real_profile_cdp()` reuse probe to use `_cdp_from_data_dir(copy_dir)` + `_cdp_http_ready(existing)` only, dropping `_cdp_on_data_dir` check (by construction hit is on our dir). Always `_agent_browser_close_session` best-effort when no live browser owns dir, rather than only when `existing` truthy — ensures stale session closed before overlay+relaunch.
- Tests `tests/tools/test_browser_real_profile.py:29-131`: replaces `side_effect=[None,"..."]` with `return_value` to avoid triggering launch, adds `test_reuse_probe_never_calls_agent_browser_get_cdp` asserting `get_cdp.call_count==1` (only post-launch attach), `test_stale_port_file_is_closed_not_reused` writes stale `DevToolsActivePort` "5000" and expects close count 1 and fresh cdp "41000", and `test_cdp_from_data_dir_reads_devtoolsactiveport`.

Non-blocking:
- `_cdp_from_data_dir` reads only first line, ignores second line (`/devtools/browser/x`) — correct per Chrome file format. **Non-blocking**.
- Probing via file read races with browser startup writing the file; `_cdp_http_ready` check mitigates stale read. **Non-blocking**.
- Existing `_cdp_on_data_dir` helper remains but no longer used in this path — consider removing dead code or keeping for other callers. **Non-blocking**.

Verdict: Correctly fixes throwaway-browser resurrection bug with solid test coverage.

---

PR #92934, opened 2026-08-23
fix(update): preserve active receipt across module purge

bluefateludi:
## Summary

- preserve `hermes_cli.update_receipt` during the post-pull stale-module purge
- add an integration regression test for begin -> purge -> finalize, including both receipt files

## Root cause

The updater begins a receipt before changing the checkout, then purges cached Hermes modules before the gateway restart phase. Because `hermes_cli.update_receipt` was not protected, the later import created a fresh module with `_current = None`, so successful updates silently skipped receipt persistence.

## Tests

- `scripts/run_tests.sh tests/hermes_cli/test_update_stale_module_purge.py tests/hermes_cli/test_update_receipt.py tests/hermes_cli/test_update_inventory.py` (39 passed)
- `python -m ruff check hermes_cli/update_cmd.py tests/hermes_cli/test_update_stale_module_purge.py`

Fixes #92535

PR #98449, opened 2026-08-30
fix(update): keep the in-flight update receipt module alive across the stale purge

### LINKED ISSUE #98436: [Bug][Windows] hermes update hand-off child silently loses its update receipt after the catch-up module purge — console end state also looks hung
### Environment

- OS: Windows 11 (10.0.26200, x64), PowerShell / Windows Terminal
- Hermes Agent v0.20.6 (2026.8.27), upstream 26350357, install method: git
- Python 3.11.15 (venv)

### Summary

On Windows, every `hermes update` that goes through the `hermes.exe` shim hand-off (parent prints *"finishing the dependency install under the venv Python… this shell returns right away"* and exits) spawns the child via `python -m hermes_cli.main update` + `HERMES_UPDATE_REEXEC=1`. When that child's run includes the pending-fleet-restart catch-up — which is the common case, because the parent writes the `fleet_restart_pending` marker right after the pull — **the child's update receipt is silently never written**. The only receipt on disk is the parent's `sys.exit(0)`, so the entire dependency-sync work (the riskiest part of a Windows update) leaves no diagnostic trace.

The same flow also produces a confusing console end state: the parent returned the shell prompt immediately, the child keeps printing into the shared console, and when the child finishes there is no completion cue and no visible prompt — the update *looks* hung. Users press Enter and re-run `hermes update` unnecessarily (I did, daily, for a week before investigating).

### Repro

1. On Windows, `git reset --hard` main back one day (or otherwise ensure new commits + the `fleet_restart_pending` marker path will trigger).
2. Run `hermes update` in a real console via the `hermes.exe` shim.
3. Parent pulls, writes the `fleet_restart_pending` marker, spawns the hand-off child, exits with receipt `"stop_reason": "sys.exit(0)"` ✔
4. Child performs the dependency sync, prints `✓ Update complete!`, runs the pending fleet restart catch-up, clears the marker, exits — **no receipt file is created for the child**. `latest.json` still points at the parent's receipt.
5. Repeat on any day: `update_receipts/` shows daily *pairs* (parent `sys.exit(0)` ≈1267 B + the user's redundant second run `completed at command boundary` ≈1285 B) and never a receipt for the child that did the real work. Consistent on this machine since 2026-08-23.

### Root cause (code + runtime trace verified)

1. The child runs the catch-up: `_run_pending_fleet_restart()` (`hermes_cli/update_cmd.py:3115`) starts with `_m()._purge_stale_hermes_modules()` (`hermes_cli/update_cmd.py:90`).
2. `_STALE_PURGE_PROTECTED` (`update_cmd.py:80`) protects only `hermes_cli`, `hermes_cli.main`, `hermes_cli.update_cmd`, `hermes_cli.hermes_logging`. **`hermes_cli.update_receipt` is not protected** → it is evicted from `sys.modules`, taking the module-level singleton `_current` (the in-flight receipt begun by `begin_update_receipt()`) with it.
3. At the command boundary, `cmd_update` does `from hermes_cli.update_receipt import finalize_pending_update_receipt` — the from-import rebuilds a **fresh module object** in which `_current is None`, so `finalize_pending_update_receipt` hits its documented exactly-once no-op ("No-op when no receipt is open") and returns without writing anything.
4. The hand-off hard-exit (`os._exit` in `cmd_update`'s finally, #93581) then terminates the process. Receipt never persisted.

Monkeypatch trace of the child process (`begin`/`finalize`/`os._exit` instrumented):

```
begin_update_receipt called
begin done, _current set: True
os._exit(0) called from: … main.py:10889 in cmd_update
```

— `finalize_update_receipt` is never invoked. Control run: the same child **without** the catch-up (no marker → no purge) finalizes normally and writes its receipt.

### Impact

- The hand-off child — which performs the actual venv rewrite on Windows — is invisible to receipt-based diagnostics; `latest.json` misreports the update as the parent's bare `sys.exit(0)`.
- Console UX: after the child's output ends, the cursor sits on its last line with no prompt redraw and no completion separator, so the update appears hung; the Enter + re-run ritual is pure confusion (and creates the spurious second receipt in the daily pairs).

### Suggested fixes

- Add `hermes_cli.update_receipt` to `_STALE_PURGE_PROTECTED`, **or**
- After the purge, re-anchor the receipt singleton into the freshly imported module, **or**
- Finalize/flush the receipt before calling `_purge_stale_hermes_modules()` on the catch-up path.
- UX nicety: have the hand-off child print a distinctive completion line (e.g. `✓ Hand-off install finished — your shell prompt was already returned`) so the frozen-looking end state is distinguishable from a real hang.

liuhao1024:
## What does this PR do?

On Windows, every `hermes update` that goes through the `hermes.exe` shim hand-off spawns a child that runs the pending-fleet-restart catch-up mid-update. `_run_pending_fleet_restart` begins with `_purge_stale_hermes_modules()`, which evicted `hermes_cli.update_receipt` from `sys.modules` along with every other `hermes_cli`-prefixed module — taking the module-level `_current` singleton (the receipt opened by `begin_update_receipt()`) with it. At the command boundary, `from hermes_cli.update_receipt import finalize_pending_update_receipt` then rebuilt a fresh module object whose `_current` is `None`, so the documented exactly-once finalize no-op'd: the child that performed the actual dependency sync (the riskiest part of a Windows update) left no receipt on disk, and `latest.json` kept pointing at the parent's bare `sys.exit(0)` record.

This adds `hermes_cli.update_receipt` to `_STALE_PURGE_PROTECTED`. The module holds cross-purge in-flight state — exactly the situation the protected set exists for (its docstring: modules whose eviction buys nothing because running state depends on them). The later from-import now binds to the same module object and the singleton survives to be finalized. Of the fix options suggested in the issue, this is the smallest and the only one that doesn't change any timing: re-anchoring the singleton or finalizing before the purge both reorder receipt bookkeeping around the catch-up.

## Related Issue

Fixes #98436

## Type of Change

## Changes Made

- `hermes_cli/update_cmd.py` — added `hermes_cli.update_receipt` to `_STALE_PURGE_PROTECTED` with a comment explaining the singleton-preservation rationale
- `tests/hermes_cli/test_update_stale_module_purge.py` — added `test_purge_preserves_inflight_update_receipt_singleton`: begins a real receipt, runs the purge, and asserts the later from-import still binds to the same module object and the command-boundary finalize persists the receipt file

## How to Test

1. `python -m pytest tests/hermes_cli/test_update_stale_module_purge.py -q` — Observed result: 6 passed
2. Negative anchor: with only the test (protection stashed), the new test fails on `sys.modules.get("hermes_cli.update_receipt") is update_receipt` — Observed result: 1 failed, 5 passed
3. `python -m ruff check hermes_cli/update_cmd.py tests/hermes_cli/test_update_stale_module_purge.py` — All checks passed

Note: `tests/hermes_cli/test_update_fleet_restart_pending.py::test_marker_written_after_pull_cleared_after_successful_restart` fails on this machine both with and without this change — a local environment artifact (a live gateway process and a `gateway_state.json` left by earlier pytest runs feed the real fleet probe), not a regression from this PR.

### Code

### DIFF
--- a/hermes_cli/update_cmd.py
+++ b/hermes_cli/update_cmd.py
@@ -77,11 +77,19 @@ _STALE_PURGE_PREFIXES = (
 #: code currently EXECUTING the update, so evicting them buys nothing — the
 #: running frames keep their module objects alive regardless — and reloading
 #: them mid-flight is the one genuinely unsafe move.
+#: ``hermes_cli.update_receipt`` is protected for the singleton's sake: it
+#: holds the in-flight receipt's module-level ``_current`` between
+#: ``begin_update_receipt()`` and the command-boundary finalize. The
+#: fleet-restart catch-up runs mid-update, so purging it there would make
+#: the later ``from hermes_cli.update_receipt import ...`` rebuild a fresh
+#: module whose ``_current`` is None and the exactly-once finalize no-op —
+#: silently losing the receipt of the very process doing the work (#98436).
 _STALE_PURGE_PROTECTED = frozenset(
     {
         "hermes_cli",
         "hermes_cli.main",
         "hermes_cli.update_cmd",
+        "hermes_cli.update_receipt",
         "hermes_cli.hermes_logging",
     }
 )
--- a/tests/hermes_cli/test_update_stale_module_purge.py
+++ b/tests/hermes_cli/test_update_stale_module_purge.py
@@ -82,6 +82,41 @@ def test_purge_protects_executing_modules():
     assert "hermes_cli" in sys.modules
 
 
+def test_purge_preserves_inflight_update_receipt_singleton(tmp_path, monkeypatch):
+    """The in-flight receipt must survive the purge fired by the catch-up (#98436).
+
+    ``begin_update_receipt`` stores the receipt in the module-level
+    ``_current`` singleton. The fleet-restart catch-up
+    (``_run_pending_fleet_restart``) purges cached Hermes modules mid-update,
+    before the command boundary runs ``from hermes_cli.update_receipt import
+    finalize_pending_update_receipt``. If the purge evicts that module, the
+    from-import rebuilds a fresh module whose ``_current`` is None and the
+    exactly-once finalize no-ops — the Windows hand-off child that did the
+    real dependency work leaves no receipt on disk.
+    """
+    from hermes_cli import update_receipt
+
+    monkeypatch.setattr(update_receipt, "_receipt_dir", lambda: tmp_path)
+
+    update_receipt.begin_update_receipt()
+    assert update_receipt._current is not None, "precondition: receipt is open"
+
+    cli_main._purge_stale_hermes_modules()
+
+    # The purge must not evict the module holding the in-flight receipt: the
+    # later from-import binds to the SAME module object, singleton intact.
+    assert sys.modules.get("hermes_cli.update_receipt") is update_receipt
+
+    from hermes_cli.update_receipt import finalize_pending_update_receipt
+
+    path = finalize_pending_update_receipt(stop_reason="cmd boundary")
+
+    assert path is not None and path.exists(), (
+        "receipt finalized after purge must be persisted"
+    )
+    assert update_receipt._current is None
+
+
 def test_purge_leaves_prefix_lookalikes_alone():
     # `gateway_foo` starts with the string prefix "gateway" but is NOT the
     # gateway package — the root-segment check must spare it.


alt-glitch:

Duplicate of #92934: both add hermes_cli.update_receipt to the stale-purge protected set and cover preserving the active receipt through finalization.

liuhao1024:
Closing this PR in favor of #92934 by @bluefateludi, which was opened a week earlier and implements the same minimal fix: adding `hermes_cli.update_receipt` to `_STALE_PURGE_PROTECTED` in `hermes_cli/update_cmd.py`, with an equivalent begin → purge → finalize regression test. I don't see anything in my version that the earlier PR doesn't already cover, so the earlier PR should take this. Thanks to the triage bot for catching the overlap.

The underlying issue (#98436) remains valid — the receipt loss through stale-module purge is real and reproducible.

PR #98452, opened 2026-08-30
feat(goals): add model-callable goal control

royalaid:
## Summary
- expose session-bound `goal_control` set/status/pause/resume/clear actions to models
- persist structured acceptance evidence and rotate read-back receipt tokens on mutation
- serialize cross-process goal mutations while allowing pause and clear to complete during slow gate or judge evaluation
- add focused control, provenance, migration, and documentation coverage

## Verification
- focused goal-control and quality-gate matrix: 45 passed
- both blocked-judge pause and blocked-gate clear real-process regressions pass
- `git diff --check upstream/main...HEAD`
- final rebase range-diff: all 10 patches unchanged

## Known upstream Windows baseline
`test_run_gate_fail_captures_output` still fails on native Windows because its POSIX compound shell command returns success there. The same failure was reproduced on the upstream base; this PR does not change `run_gate`.

Head: `6286a22632b74a7215b4ca970c3c3002490ab010`


andrexibiza (commented):
Reviewed the exact current object `6286a22632b74a7215b4ca970c3c3002490ab010` against its exact merge base/current `main@2a598aad1c398e95b3325a0f100f5c28efa63d12` (`10 ahead / 0 behind`). I read all seven changed files, the focused tests, the current GoalManager persistence/evaluation paths, the #24477 predecessor, the still-open #98331 competing/successor branch, and the existing review history. Exact-head hosted acceptance is green: CI `33302308779`, Docker `33302308450`, and Nix `33302308447` all succeeded; the CI run includes Python, e2e, Windows/macOS, blocking ruff/Windows-footguns, attribution, and a green `All required checks pass` job.

There is strong work here. The cross-process file lock is scoped by profile+session rather than globally; the slow judge/gate work is deliberately outside the mutation lock; pause/clear are exercised from real spawned processes while evaluation is blocked; canonical post-mutation readback is compared against the intended state; and the seven-file rewrite is substantially tighter than the 19-file shape in #98331. Those are meaningful improvements.

I still see three authority/state blockers before this can be the convergence object.

### P1 — `goal_control` can mint fresh autonomy budget and resurrect terminal goals

The model-facing boundary is currently strictly weaker than both the existing #24477 safety work and the current #98331 contract.

In `tools/goal_control_tool.py`, `set` and `update` are routed to the same `manager.set(...)`. `GoalManager.set()` constructs a fresh `GoalState` with `turns_used=0`, and the tool accepts any positive `max_turns`; there is no cap at configured `goals.max_turns`. The new regression `test_update_replaces_condition_and_budget` explicitly blesses `max_turns: 2 -> 9` rather than preserving the bounded-run invariant. A model can therefore repeatedly call `update` (or `set`) to reset consumed turns and enlarge its own continuation budget. `resume` has the same problem: the tool calls `manager.resume()` without `reset_budget=False`, while `GoalManager.resume()` defaults `reset_budget=True` and zeros `turns_used`.

There is also a terminal-state composition hole. The wrapper rejects a direct `resume` of a cleared row, but `pause` accepts every non-paused persisted status. `GoalManager.pause()` unconditionally rewrites the row to `paused`. So `clear -> pause -> resume` (and likewise `done -> pause -> resume`) turns terminal state back into an active goal, with the resume resetting consumed turns.

This is the exact "other side of the shape" that action-by-action tests miss: each individual transition looks plausible, but legal composition creates new authority.

**Required fix:** make the model budget monotonic and terminal statuses absorbing. `set` must not overwrite a live persisted goal; `update` must be limited to active/paused state, preserve `turns_used` and paused status, and never increase the effective budget; any supplied budget must be bounded by configured `goals.max_turns`; model `resume` must preserve consumed turns; `pause`/`resume`/`update` must reject `done`/`cleared` (and exhausted terminal/paused states according to the canonical policy) rather than rewriting them. Please pin the adversarial compositions directly: consumed-turn update, oversized set/update, exhausted pause/resume, `clear -> pause -> resume`, and `done -> pause -> resume`.

This is not speculative precedent: #24477 already has explicit `max_turns`-above-configured rejection coverage, and #98331's current contract already states that update cannot enlarge budget, resume preserves consumed turns, and terminal goals cannot be reactivated. Preserve those safety properties in the tighter implementation rather than regressing them.

### P1 — the migration is serialized, but it is not atomic and can report success after a partial/dropped write

`migrate_goal_to_session()` now acquires both session locks in deterministic order, which fixes writer interleaving. But after the locks are held it still performs two independent persistence operations:

1. `save_goal(new_session_id, child_state)`
2. mutate the parent to `cleared`, then `save_goal(old_session_id, state)`

`save_goal()` catches `set_meta` exceptions and returns no success/failure signal. That means a child-write failure can be silently followed by a parent clear, losing the active goal; a parent-write failure can leave both parent and child active; and either failure can still fall through to `migrate_goal_to_session()` returning `True`. A process death between the two writes produces the same split-brain state even though the file locks were correct.

The new migration test proves only that migration waits for the child-session lock. It does not prove all-or-nothing state transfer. The current #98331 branch had already moved this class toward a single durable CAS/transaction with rollback semantics; that property needs to survive the simplification.

**Required fix:** perform the parent+child transfer inside one durable SessionDB transaction/atomic compare-and-set boundary and make persistence failure observable to the caller. Add fault-injection at both write points plus `BaseException`/process-interruption coverage proving the only externally visible outcomes are pre-migration or fully migrated — never two active rows, no active row, or `True` after a dropped write.

### P1 — default-on model mutation authority is still widening past the unresolved predecessor boundary

`toolsets.py` adds `goal_control` to `_HERMES_CORE_TOOLS`, and `test_goal_control_is_exposed_on_the_core_model_surface` intentionally pins that default exposure. That is the same unresolved policy boundary called out on #24477: its current port explicitly keeps `set_goal` out of `_HERMES_CORE_TOOLS` / `_AGENT_LOOP_TOOLS` until an upstream decision is made because core membership propagates the schema through the composed `hermes-*` surfaces. The relevant predecessor review is https://github.com/NousResearch/hermes-agent/pull/24477#issuecomment-4954271462.

This implementation grants more authority than #24477 did — rewrite, pause, resume, clear, plus acceptance-evidence persistence — so the default-on decision cannot be smuggled in as a registration detail. The dedicated `goal` toolset already exists and is the natural explicit authority owner.

**Required fix:** absent an explicit repository decision that can be cited, keep `goal_control` opt-in under the `goal` toolset and remove it from `_HERMES_CORE_TOOLS`. If default-on is deliberately accepted, make that a first-class PR contract and prove exactly which CLI/cron/messaging surfaces receive the mutation authority and why that exposure is intended.

### Convergence, provenance, and landing topology

#98452 and #98331 are not independent features that should both land. They are same-author, same-surface competing implementations of the same model goal-control owner. #98452 has the better current-main composition and much smaller file footprint; #98331 currently carries the stronger bounded-budget/terminal/CAS semantics and explicit origin chain. Treat #98452 as the **superseding convergence candidate** only if it absorbs those stronger invariants, then close/supersede the overlapping #98331 path rather than maintaining two authorities.

The origin chain also needs to stay visible. #24477 / Fewmanism is the direct implementation precursor, and its budget-cap/default-opt-in work materially constrains this surface. #98331 explicitly preserved that authored precursor and documented it; #98452 currently has ten Royalaid-authored commits and no corresponding prior-work/supersession section. Preserve #24477's contributor credit and the design lineage already documented on #98331 instead of flattening the rewrite into a fresh origin. #29830 and #53268 remain historical/adjacent model-goal attempts; they are prior art, not reasons to claim unrelated scope as superseded.

Two repository landing gates remain even after the correctness repairs. First, `cli.py` is still being modified around line 13,129; this feature should not add another lifecycle seam to that godfile. Keep the refresh/ownership behavior in a bounded goal/session owner and remove the new `cli.py` accretion. Second, this submitted object is a ten-commit chain. The exact current head is fully green, but a green final head does not establish hosted acceptance for every intermediate submitted commit; retain/produce the per-commit green receipts required by this repository's landing discipline.

The cross-process pause/clear work is worth keeping. Once the model cannot expand its own budget, terminal states stay terminal, migration becomes genuinely atomic, and the exposure/provenance owners are reconciled, this can be a much cleaner convergence than the earlier broad branch.

Issue #93581, opened 2026-08-24
[Bug]: Windows: handed-off dependency install finishes (receipt exit 0, "Update complete!") but the console never returns to the shell


PaoloBarts:
## Bug Description

On Windows, `hermes update` run from the `hermes.exe` console shim now hands the dependency install to a child under the venv Python (the self-lock fix shipped in PR #90192). The hand-off message promises:

```
→ Windows: hermes.exe cannot replace itself while it runs; finishing the dependency install under the venv Python.
  The code update is already applied. The install continues below and this shell returns right away.
```

The child **completes all update work successfully** — `✓ Update complete! [main @ f293e7206b]` is printed, and the update receipt (`logs/update_receipts/update_20260823_220352_39376.json`) records `"outcome": "success", "exit_code": 0, "stop_reason": "completed at command boundary"` roughly 35 seconds after start — **but the PowerShell window never regains a prompt**. It sits frozen after the final output line for as long as it was left (~15 minutes in my case) until manually interrupted with Ctrl+C.

The update itself is fine and durable; only process exit / shell-control return is broken. This makes scripted/cron-driven updates unreliable and makes the interactive experience look hung even though nothing is wrong.

Related but distinct:
- #79040 — same failure shape (update completes, process never exits) but on Linux after "Web UI built"; mine is the Windows hand-off path introduced by #90192.
- #91675 — post-update gateway respawn silently failing (I likely hit this too in the same session, but I'm keeping this report to one bug per the maintainers' triage preference; happy to file the respawn one separately).
- #90455 — Windows Desktop hand-off hanging forever on a resident gateway (windows.ps1 `ReadToEndAsync` never sees EOF); different mechanism and surface (Desktop vs CLI), but same "hand-off completes the work, control never returns" family.
- #89599 / PR #90192 — the self-lock fix whose hand-off path produces this hang.

## Steps to Reproduce

1. On Windows 11, from PowerShell, with a pending upstream update, run:

```powershell
hermes update
```

2. Observe the hand-off banner, then the second update instance finishing:

```text
=== hermes update started 2026-08-23T22:03:03 ===   (parent, from hermes.exe shim)
→ Fleet: 1 running service(s) across profiles: default
→ Stopping Windows gateway process(es) before updating Hermes...
→ Fetching updates...
→ Found 61 new commit(s)
→ Pulling updates...
→ Windows: hermes.exe cannot replace itself while it runs; finishing the dependency install under the venv Python.
[...]

PR #98454, opened 2026-08-30
fix(update): protect hermes_cli.update_receipt from the stale-module purge (#98436)

### LINKED ISSUE #98436: [Bug][Windows] hermes update hand-off child silently loses its update receipt after the catch-up module purge — console end state also looks hung

lEWFkRAD:
## What does this PR do?

Fixes #98436: on Windows, every `hermes update` that goes through the `hermes.exe` shim hand-off loses the hand-off **child's** update receipt.

`cmd_update` begins the receipt early via `begin_update_receipt()`; the receipt lives in a module-level singleton (`_current`) inside `hermes_cli/update_receipt.py`. Later, `_run_pending_fleet_restart()` (the catch-up at `update_cmd.py:3123`) calls `_purge_stale_hermes_modules()` — and `hermes_cli.update_receipt` was **not** in `_STALE_PURGE_PROTECTED`, so the purge evicted it from `sys.modules`. The command-boundary safety net then runs `from hermes_cli.update_receipt import finalize_pending_update_receipt`, which rebuilds a **fresh module object** with `_current is None`; `finalize_pending_update_receipt` hits its documented no-op ("No-op when no receipt is open"), and the hand-off child's `os._exit(0)` path (#93581) terminates the process. Net: the process that performs the actual venv/dependency rewrite writes **no receipt**, and `latest.json` misreports the update as the parent's bare `sys.exit(0)`.

The fix adds `hermes_cli.update_receipt` to the protected set. The accurate invariant for that set — now spelled out in the comment — is "module holds live singleton state the executing update depends on": purging such a module strands its state in an orphaned module object, which is exactly this failure.

## Related Issue

Fixes #98436

## Type of Change

## Changes Made

- `hermes_cli/update_cmd.py` — add `"hermes_cli.update_receipt"` to `_STALE_PURGE_PROTECTED`, with a comment stating the mechanism and the accurate protection rationale (live singleton state, not "currently executing code").
- `tests/hermes_cli/test_update_receipt_purge_98436.py` — new regression tests:
  - `test_purge_protects_update_receipt_module` — the module must survive the purge (RED on main, GREEN on branch).
  - `test_boundary_from_import_finalizes_after_purge` — drives the exact production shape: `begin_update_receipt()` → `_purge_stale_hermes_modules()` → `from hermes_cli.update_receipt import finalize_pending_update_receipt` → assert the receipt file is actually written. This is the boundary-safety-net path from `cmd_update`, not just a frozenset-membership assertion.

## How to Test

1. On `main` (before the fix), run the new tests: `pytest tests/hermes_cli/test_update_receipt_purge_98436.py -q` → both fail (purge evicts the module; boundary from-import finalize returns `None`).
2. On this branch: both pass, plus the existing purge suite `tests/hermes_cli/test_update_stale_module_purge.py` stays green (5 tests) — protection of the receipt module does not weaken the class-fix for the 2026-08-20 stale-`cli_output` field failure.
3. Blast radius: `hermes_cli.update_receipt` imports only stdlib plus fully-lazy siblings (`build_info`, `config`, `gateway.status` — the latter re-imports fresh after the purge inside fleet verification), and it is never reloaded — only exempted from eviction — so the purge's own "reloading-in-place is the unsafe operation" invariant is untouched.
4. `scripts/check-windows-footguns.py --diff origin/main` → clean.

Tested on Windows 11 (10.0.26200, x64), Python 3.11.15.

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate (no open PR references #98436)
- [x] I've run `pytest tests/ -q` and all tests pass (ran the update-receipt + purge + fleet-restart suites: 58 passed / 27 skipped; the one failing test in `test_update_fleet_restart_pending.py` fails identically on clean `origin/main` tip — pre-existing, unrelated)
- [x] I've tested on my platform: Windows 11 (10.0.26200, x64)

### DIFF
--- a/hermes_cli/update_cmd.py
+++ b/hermes_cli/update_cmd.py
@@ -83,6 +83,18 @@ _STALE_PURGE_PROTECTED = frozenset(
         "hermes_cli.main",
         "hermes_cli.update_cmd",
         "hermes_cli.hermes_logging",
+        # (#98436) The in-flight update receipt is a module-level singleton
+        # (_current) in hermes_cli.update_receipt — the module holds live
+        # state the executing update depends on, so its cache entry must
+        # survive. Purging it strands the receipt in an orphaned module
+        # object: the next `from hermes_cli.update_receipt import ...`
+        # (cmd_update's command-boundary safety net) rebuilds a FRESH
+        # module with ``_current is None``, the finalize hits its
+        # documented no-op, and the run that did the real work — the
+        # Windows hand-off child, whose pending fleet-restart catch-up
+        # purges here (_run_pending_fleet_restart) — silently writes NO
+        # receipt.
+        "hermes_cli.update_receipt",
     }
 )
 
--- a/tests/hermes_cli/test_update_receipt_purge_98436.py
+++ b/tests/hermes_cli/test_update_receipt_purge_98436.py
@@ -1,0 +1,89 @@
+"""Regression tests for #98436 — update receipt lost when the stale-module
+purge evicts hermes_cli.update_receipt mid-run.
+
+Windows `hermes update` hand-off child: the pending fleet-restart catch-up
+starts with _purge_stale_hermes_modules(), which evicted the cached
+hermes_cli.update_receipt module (it was not in _STALE_PURGE_PROTECTED).
+The singleton receipt lives in that module, so the later
+`from hermes_cli.update_receipt import finalize_pending_update_receipt`
+in cmd_update's command-boundary safety net re-imported a FRESH module
+with `_current is None` -> finalize hit its documented no-op -> the
+hand-off child (the process that does the real dependency sync) wrote NO
+receipt.
+
+Fix: protect hermes_cli.update_receipt from the purge. Accurate invariant
+for the protected set here: the module holds LIVE SINGLETON STATE that the
+currently-executing update depends on — evicting its cache entry strands
+that state in an orphaned module object.
+"""
+
+from __future__ import annotations
+
+import sys
+
+import pytest
+
+from hermes_cli import main as cli_main
+
+
+@pytest.fixture(autouse=True)
+def _restore_sys_modules():
+    snapshot = dict(sys.modules)
+    yield
+    for name, mod in snapshot.items():
+        sys.modules[name] = mod
+    for name in list(sys.modules):
+        if name not in snapshot:
+            del sys.modules[name]
+
+
+def test_purge_protects_update_receipt_module():
+    """#98436: the receipt singleton lives in hermes_cli.update_receipt.
+
+    Evicting it mid-update silently drops the in-flight receipt: a later
+    from-import rebuilds the module with `_current is None` and finalize
+    becomes a no-op, so the run that did the real work leaves no receipt.
+    """
+    import hermes_cli.update_receipt as update_receipt
+
+    cli_main._purge_stale_hermes_modules()
+
+    assert sys.modules.get("hermes_cli.update_receipt") is update_receipt, (
+        "hermes_cli.update_receipt was purged — the in-flight receipt "
+        "singleton dies with it (#98436)"
+    )
+
+
+def test_boundary_from_import_finalizes_after_purge(tmp_path, monkeypatch):
+    """Drive the EXACT failure shape from cmd_update's boundary safety net:
+
+        begin -> purge -> `from hermes_cli.update_receipt import
+        finalize_pending_update_receipt` -> call it
+
+    asserting the receipt is actually written. On main this is RED: the
+    purge evicts the module, the from-import rebuilds it with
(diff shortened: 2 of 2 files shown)


alt-glitch:

Duplicate of #92934: both protect the active update_receipt module from the stale-module purge, preserving the same receipt-finalization mechanism.

teknium1:
Same fix landed in #103478 (f58fcc8118) via the earlier #92934 by @bluefateludi (cherry-picked with credit). Your writeup of the Windows hand-off child path was useful confirmation — thanks.

Issue #57491, opened 2026-07-03
Rotation compression: _db_persisted marker copied via messages[i].copy() skips bulk flush to child session (cached gateway regression since #50372)


hejuntt1014:
## Summary

When context compression rotates to a **child session** (`compression.in_place: false`, the legacy/default rotation path), the compressed transcript (~1000+ messages including the LLM summary) is **not persisted** to the new session in `state.db`. Only messages appended **after** compression (new dicts without `_db_persisted`) are written incrementally.

The live gateway cached agent keeps the full compressed context in `_session_messages`, so the agent continues working until restart/eviction. After restart, `load_transcript()` returns the nearly-empty child session → **severe amnesia** (summary + tail lost from durable storage).

This is a **regression** introduced by the intrinsic `_db_persisted` flush dedup (#50372, commit `e4c6d1b22`, merged 2026-07-01). It is **not** the same root cause as the earlier fixes for stale `conversation_history` (#43062) or flush-before-rotate to the parent (#48584 / #47202).

## Root cause

1. **`_flush_messages_to_session_db`** skips any message dict that already carries `_db_persisted` (`run_agent.py`):

   ```python
   if msg.get(_DB_PERSISTED_MARKER):
       continue
   ```

2. **`ContextCompressor.compress()`** builds the post-compression list with **`messages[i].copy()`** for protected head and tail (`context_compressor.py`). Shallow `.copy()` **propagates** `_db_persisted: True` from the live cached-agent transcript.

3. On a **long-lived gateway cached agent**, incremental flushes stamp `_db_persisted` on essentially every message in `_session_messages` over many turns.

4. After rotation, `conversation_history_after_compression()` correctly returns `None`, and `_last_flushed_db_idx` is reset — but the marker skip still drops **all** copied head/tail rows when flushing to the **new** `session_id`.

5. **Standalone summary messages** are new dicts and should flush — unless `_merge_summary_into_tail` embeds the summary into the first tail message (which carries `_db_persisted`), in which case the summary is lost from DB as well.

**Before #50372:** dedup used `{id(msg)}`; `.copy()` produced new dicts with new `id()` values, so rotation flush to the child worked.

**After #50372:** `.copy()` + marker = child session gets only post-compression appends.

## Production evidence (WeChat gateway, rotation mode)

Profile: long-running cached gateway session, `compression.in_place: false`.

| Event | Time (UTC+8) | Observation |
|-------|----------------|-------------|
[...]

PR #97006, opened 2026-08-28
fix(compression): use the route-aware pruned estimate in the mid-turn pre-API guard

liuhao1024:
## What does this PR do?

Aligns the mid-turn pre-API compression guard with the turn-prologue preflight so both measure the same route-aware wire payload on native-Codex sessions.

The #96155 fix (#96644) made the **turn-prologue** preflight estimate the checkpoint-pruned native Responses payload. The independent **mid-turn** guard in `conversation_loop` still estimated the full assembled durable history, so on a compacted native-Codex session it false-triggered a 600-second local compression the main request never needed: the issue's deterministic, credential-free probe on clean main shows the generic figure at 1,037,241 tokens versus 6,036 pruned (171x), and the live sequence shows the actual request then fitting comfortably (164k input against a 765k threshold) after the wasted ten-minute compression (#96995).

The guard's pressure figure is extracted into `_midturn_request_pressure_tokens`, which mirrors the turn-prologue: when native Responses compaction is proven eligible it uses `estimate_native_responses_preflight_tokens` (checkpoint-pruned, system prompt and tool schemas included); otherwise it keeps the generic message+tools figure. Passing the assembled `api_messages` (which carry the system row) alongside `effective_system` counts the system prompt exactly once — the estimator's converter skips system-role rows and adds the prompt separately. `total_chars` (a verbose-log proxy) and every non-codex path are unchanged.

## Related Issue

Fixes #96995

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `agent/conversation_loop.py` — new module-level `_midturn_request_pressure_tokens(agent, api_messages, effective_system, approx_tokens)`; the pre-API guard now computes `request_pressure_tokens` through it instead of the unconditional generic `estimate + tools` figure.
- `tests/agent/test_native_preflight_estimate.py` — two regressions mirroring the prologue tests: eligible route returns the pruned estimate against the assembled (system-carrying) messages without double-counting the system prompt; ineligible route falls back to generic + tools exactly as before.

## How to Test

1. `python -m pytest tests/agent/test_native_preflight_estimate.py -q` — should pass (8 passed, including the two new regressions).
2. Red/green verified locally: with the `conversation_loop.py` change reverted, the two new mid-turn tests fail (helper absent); with this change all pass.
3. Adjacent suites unchanged: `tests/agent/test_compress_context_progress_timeout.py`, `test_compress_focus.py`, `test_compress_signal_leak.py` — 18 passed.
[...]

PR #98458, opened 2026-08-30
fix(compression): stamp _DB_PERSISTED_MARKER after in-place batch commit

### LINKED ISSUE #98450: fix(compression): in-place compaction commit never stamps _DB_PERSISTED_MARKER — post-compaction transcript re-INSERTs, live set regrows ~58K → ~512K tokens
## Summary

With `compression.in_place: true`, the batch compaction commit writes the compacted set through `SessionDB.archive_and_compact()` but **never stamps `_DB_PERSISTED_MARKER` on the compacted dicts**. `compress()` returns marker-swept COPIES (`_strip_persistence_markers`), so the next `_persist_session` → `_flush_messages_to_session_db_unlocked` walk sees every compacted row as unpersisted and **re-INSERTs the whole post-compaction transcript**. The micro-compaction path (`ContextCompressor._sync_micro_compact_to_db`, context_compressor.py) does stamp the marker after its `archive_and_compact` call — the in-place batch path simply lacks the same post-commit contract.

## Environment

- Hermes Agent v0.20.6 baseline + local production deployment (branch `upgrade/v0206-minimal-20260829`, base `5fc308a707…`); also verified the invariant is still unmet on `origin/main` at `26350357d7` (checked `HEAD..origin/main`, 552 commits incl. the #94996 rotation-dedupe salvage and attempt-ownership fixes — those fix a different duplicate-write site and the compressor state races, not this path).
- Models: gpt-5.6-sol via openai-codex (coordinator), k3 via kimi-coding (auditor1) — model-independent; it is a persistence-layer bug.

## Production evidence (2026-08-30)

Coordinator session `20260830_121928_330881ba`:

- A 223,746-token preflight compression committed in place at 13:54 (`split_status: in_place_committed`, 46→29 messages, ~58,614 tokens).
- The **same batch was durable-written twice with byte-identical timestamps** at commit: state.db rows `529239` and `529304` share timestamp `1788069248.6359262` and tool_call_id `call_NqMk…` (73,690-char skill_view result). It was written a **third** time at 14:19 by the micro-compaction transcript-recovery pass (61 rows re-inserted, `compacted=1`), and a **fourth** time at 15:02 at turn finalize (row `529692`, `active=1`).
- Live set regrew ~58K → ~487K → ~512K tokens within two hours; every later preflight compression timed out at the 600 s ceiling ("Context compression made no progress for 0.0s (total wait 600.0s…)") because the session no longer fit the 272K model window. Repeated `/new` on the *old* session appeared to do nothing because each regeneration hit the same runaway transcript.
- Auditor1 session `20260829_205018_6a336f35` showed the same class at greater scale: **41 duplicate content groups, ~3.0 MB redundant** (one 55,751-char terminal output present ×6), ~595K tokens live, same repeated 600 s compression-timeout failures.

The API provider was ruled out: replaying the exact 1.28M-char coordinator payload to the auxiliary summarizer completed in 45.3 s; the 525K-token auditor1 payload completed in 12.8 s. The failures are purely the local transcript growth + retry loop.

## Root cause

`run_agent.py::_flush_messages_to_session_db_unlocked` skips any dict carrying `_DB_PERSISTED_MARKER` (the #860 dedupe contract). Messages that have *not* been persisted are written. After an in-place commit:

1. `compress()` returns copies whose markers were stripped by design (`_strip_persistence_markers`, needed so the rotation flush doesn't skip compacted rows — #57491);
2. the in-place branch commits them via `archive_and_compact(...)` and resets the legacy `agent._flushed_db_message_ids` seed — but stamps nothing;
3. the next persist (turn finalize, close safety-net, or mid-turn flush) re-scans and re-writes the entire compacted set as "new";
4. the next preflight compression then runs over the doubled transcript, commits, and gets doubled again — monotonic growth.

The comment at conversation_compression.py:3676 asserts the compacted dicts are "passed as conversation_history next turn and skipped by identity" — but identity-skip only works for dicts that stay in `conversation_history`; dicts handed back as the live `messages` list are marker-swept copies and are re-written.

## Minimal fix (13 lines, in the `if in_place:` branch right after `split_status = "in_place_committed"`)

```python
# Stamp the persistence marker on the compacted dicts — same post-commit
# contract as ContextCompressor._sync_micro_compact_to_db. compress()
# returns marker-swept COPIES; without this stamp the next
# _persist_session re-INSERTs the whole post-compaction transcript.
from agent.context_compressor import _DB_PERSISTED_MARKER
for _compacted_msg in compressed:
    if isinstance(_compacted_msg, dict):
        _compacted_msg[_DB_PERSISTED_MARKER] = True
```

This mirrors the existing stamp in `_sync_micro_compact_to_db` and is safe for the flush's seed/identity paths (markers on the live dicts are exactly what the flush contract expects; `_strip_persistence_markers` only guarantees the *compress()* output invariant).

## Regression test

`TestInPlaceCommitStampsPersistenceMarkers` (added to `tests/run_agent/test_in_place_compaction.py`):

- `test_committed_dicts_carry_marker` — asserts every dict returned by `compress_context` carries the marker after an in-place commit;
- `test_no_duplicate_rows_after_post_commit_persist` — reproduces the production sequence (in-place commit → `_persist_session` over the compacted live list) and asserts one active row per compacted message, zero duplicate contents, `8 + len(compressed)` total rows.

With the fix reverted (git stash), the new tests fail reproducing the bug; with the fix, the full file passes 13/13 and the neighboring compression-persistence suites (29 + 18 tests) stay green.

## Request

Upstream fix in `agent/conversation_compression.py`'s in-place commit branch (or, alternatively, making `archive_and_compact()`'s caller contract explicit and stamping inside a shared helper used by both batch and micro paths). Happy to open a PR with the patch + tests if the direction looks right.

liuhao1024:
## What does this PR do?

The in-place batch compaction commit writes the compacted set through `SessionDB.archive_and_compact()` but never stamps `_DB_PERSISTED_MARKER` on the compacted dicts. `compress()` returns marker-swept copies (#57491 invariant) that come back as the live `messages` list, while the next `_persist_session` → `_flush_messages_to_session_db_unlocked` walk still receives the pre-compaction dicts as `conversation_history` — the flush's identity skip cannot cover the compacted dicts, so every compacted row looks unpersisted and the **entire post-compaction transcript is re-INSERTed** as duplicate active rows. Each subsequent preflight compression then runs over the doubled transcript and doubles it again (the monotonic live-set growth in #98450, until every compression hits the 600s timeout).

The fix mirrors the existing post-commit contract of the micro-compaction path (`ContextCompressor._sync_micro_compact_to_db`): after a successful `archive_and_compact()`, stamp `_DB_PERSISTED_MARKER = True` on every committed dict so the next append-only flush skips them. On failure the sweep stays, matching the micro path's rollback semantics.

## Related Issue

Fixes #98450

## Type of Change

## Changes Made

- `agent/conversation_compression.py` — in the `if in_place:` commit branch, extend the existing local import with `_DB_PERSISTED_MARKER` and stamp it on every compacted dict after `archive_and_compact()` succeeds; rewrote the adjacent `_flushed_db_message_ids` comment whose identity-skip rationale was the blind spot this fix closes
- `tests/run_agent/test_in_place_compaction.py` — added `TestInPlaceCommitStampsPersistenceMarkers`: one test asserting the committed dicts carry the marker, one replaying the production sequence (in-place commit → flush over the compacted live list with the pre-compaction history) and asserting one active row per compacted message with zero duplicate contents

## How to Test

1. `pytest tests/run_agent/test_in_place_compaction.py -q` — 13 passed (11 existing + 2 new)
2. Negative anchor: with the fix stashed, both new tests fail reproducing the bug (missing marker; 4 active rows with duplicated contents instead of 2); with the fix, all pass — Observed result: 2 failed without the fix, 13 passed with it
3. Neighboring compression-persistence suites stay green: `pytest tests/run_agent/test_compression_persistence.py tests/agent/test_compressed_summary_metadata.py tests/run_agent/test_413_compression.py tests/agent/test_micro_compaction.py tests/run_agent/test_compression_closed_adoption.py -q` — 92 passed
4. `ruff check agent/conversation_compression.py tests/run_agent/test_in_place_compaction.py` — clean

### Code

- [x] I've tested on my platform: macOS 15.2 (arm64)

## For New Skills

N/A

### DIFF
--- a/agent/conversation_compression.py
+++ b/agent/conversation_compression.py
@@ -4156,6 +4156,7 @@ def compress_context(
                     # WITHOUT destroying history, unlike a hard replace_messages).
                     # See #38763.
                     from agent.context_compressor import (
+                        _DB_PERSISTED_MARKER,
                         PROACTIVE_PRUNE_REARM_MODEL_CONFIG_KEY,
                     )
 
@@ -4169,11 +4170,24 @@ def compress_context(
                         lock_holder=_lock_holder,
                     )
                     split_status = "in_place_committed"
-                    # Reset the flush identity set so the next turn's appends are
-                    # diffed against the COMPACTED transcript: the compacted dicts
-                    # are passed as conversation_history next turn and skipped by
-                    # identity, so only genuinely new turn messages get appended
-                    # (no dup of the summary, no resurrection of dropped turns).
+                    # Stamp the persistence marker on the committed dicts —
+                    # the same post-commit contract as the micro-compaction
+                    # sync (_sync_micro_compact_to_db). compress() returns
+                    # marker-swept copies (#57491), and the compacted dicts
+                    # come back as the live ``messages`` list while the flush
+                    # still receives the pre-compaction dicts as
+                    # ``conversation_history``, so its identity skip cannot
+                    # cover them: without this stamp the next _persist_session
+                    # re-INSERTs the whole post-compaction transcript (#98450).
+                    # Stamped only after a successful archive_and_compact —
+                    # on failure the sweep stays, matching the micro path's
+                    # rollback semantics.
+                    for _compacted_msg in compressed:
+                        if isinstance(_compacted_msg, dict):
+                            _compacted_msg[_DB_PERSISTED_MARKER] = True
+                    # Reset the flush identity set: stale ids from the
+                    # pre-compaction dicts must not survive into the next
+                    # flush's one-shot seed translation.
                     agent._flushed_db_message_ids = set()
                     # Rotation-independent signal: the conversation was compacted in
                     # place (id unchanged). The gateway reads this (NOT an id-change
--- a/tests/run_agent/test_in_place_compaction.py
+++ b/tests/run_agent/test_in_place_compaction.py
@@ -436,3 +436,65 @@ class TestCompactedTurnsStaySearchable:
                 "ZEBRAWORD", role_filter=["user", "assistant"], include_inactive=True
             )
             assert len(recovered) == 1
+
+
+class TestInPlaceCommitStampsPersistenceMarkers:
+    """The in-place batch commit must stamp ``_DB_PERSISTED_MARKER`` on the
+    compacted dicts — the same post-commit contract as the micro-compaction
+    sync. ``compress()`` returns marker-swept copies (#57491) that come back
+    as the live ``messages`` list, while the flush still receives the
+    pre-compaction dicts as ``conversation_history``; without the stamp the
+    identity skip cannot cover them and the next persist re-INSERTs the whole
+    post-compaction transcript (#98450)."""
+
+    def test_committed_dicts_carry_marker(self):
+        from hermes_state import SessionDB
+        from agent.conversation_compression import compress_context
+        from agent.context_compressor import _DB_PERSISTED_MARKER
+
+        with tempfile.TemporaryDirectory() as tmp:
+            db = SessionDB(db_path=Path(tmp) / "t.db")
+            sid = "20260830_marker"
+            _seed(db, sid, "marker")
+            agent = _make_agent(db, sid, in_place=True)
+
+            messages = [{"role": "user", "content": f"m{i}"} for i in range(8)]
+            compressed, _sp = compress_context(
+                agent, messages, approx_tokens=100_000, system_message="sys"
+            )
+
+            assert agent._last_compaction_in_place is True
+            for msg in compressed:
+                assert msg.get(_DB_PERSISTED_MARKER) is True
+
+    def test_no_duplicate_rows_after_post_commit_persist(self):
+        from hermes_state import SessionDB
+        from agent.conversation_compression import compress_context
+
+        with tempfile.TemporaryDirectory() as tmp:
+            db = SessionDB(db_path=Path(tmp) / "t.db")
+            sid = "20260830_dupes"
+            _seed(db, sid, "dupes")
+            agent = _make_agent(db, sid, in_place=True)
+
+            history = [{"role": "user", "content": f"m{i}"} for i in range(8)]
+            compressed, _sp = compress_context(
+                agent, history, approx_tokens=100_000, system_message="sys"
+            )
+
+            # Production sequence (#98450): the turn-finalize persist runs over
+            # the compacted live list while conversation_history still holds the
+            # pre-compaction dicts — identity cannot dedupe these.
+            agent._flush_messages_to_session_db(compressed, history)
+
+            active = db.get_messages_as_conversation(sid)
+            assert len(active) == 2
+            assert [m.get("content") for m in active] == [
+                "[CONTEXT COMPACTION] summary of prior turns",
+                "recent reply",
+            ]
+            all_rows = db.get_messages(sid, include_inactive=True)
+            # 8 archived originals + 2 compacted actives — nothing re-inserted.
+            assert len(all_rows) == 10
+            contents = [m["content"] for m in all_rows]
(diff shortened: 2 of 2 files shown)


Enough1122:

**98458 — Stamp `_DB_PERSISTED_MARKER` after in-place compaction commit**

- `agent/conversation_compression.py:9,34-40` imports `_DB_PERSISTED_MARKER` and after successful `archive_and_compact` loops `compressed` dicts stamping `msg[_DB_PERSISTED_MARKER]=True`. Comment ties to #98450: `compress()` returns marker-swept copies (#57491) that become live `messages`, while flush still receives pre-compaction dicts as `conversation_history`; identity skip can't cover them, so next `_persist_session` re-INSERTs whole post-compaction transcript. Stamping only after success preserves micro-compaction rollback semantics. Also resets `_flushed_db_message_ids=set()` — stale ids must not survive.
- Tests `tests/run_agent/test_in_place_compaction.py:53-112` add `TestInPlaceCommitStampsPersistenceMarkers` with two cases: `test_committed_dicts_carry_marker` asserts every `compressed` dict has marker, and `test_no_duplicate_rows_after_post_commit_persist` replays production sequence `compress_context` → `_flush_messages_to_session_db(compressed, history)` and asserts DB has 8 archived + 2 active =10 rows, no dup contents, `active` correctly `["[CONTEXT COMPACTION] summary...", "recent reply"]`.

Non-blocking:
- Stamping mutates `compressed` dicts in place — safe as they are newly created by `compress_context`. **Non-blocking**.
- Ensure `_DB_PERSISTED_MARKER` key is excluded from transcript export/serialization elsewhere (typically filtered). **Non-blocking**.

Verdict: Fixes duplicate-row re-insertion after in-place compaction with precise regression tests.

---

liuhao1024:
Thanks for the review! Confirming the ordering is deliberate: the marker is stamped only after `archive_and_compact` succeeds, so a rolled-back micro-compaction never carries `_DB_PERSISTED_MARKER` into the live `messages`; and `_flushed_db_message_ids` is reset in the same block because pre-compaction ids are stale against the post-compaction transcript and must not suppress the next `_persist_session`.

andrexibiza (commented):
Reviewed exact head `fc76fb1eff5ababad9772eb1d07fe0a9f2aa136e` against the durable-write boundary, the existing marker-skip contract, current `main@4f22543509d1b91dc45bcb369447126c5eb14fb7`, and the regression witnesses.

The fix is at the right settlement boundary: `_DB_PERSISTED_MARKER` is stamped only after `archive_and_compact()` returns successfully, so failed commits do not acquire false durability state. That makes the live compacted dicts proof-carrying before the next append-only flush while preserving the existing rollback behavior. The regression covers the actual production composition—pre-compaction `conversation_history` plus post-compaction live `compressed` messages—and proves the flush cannot reinsert the committed active set.

Current-main cross-check is clean: the PR is one commit / two files, mergeable, and the 29 main-only commits since merge base do not touch either submitted path. Every surviving commit is green because there is exactly one: CI `33302615872`, Docker `33302615515`, and Nix `33302615527` all completed successfully on this exact SHA.

No blocking finding from this review.

teknium1:
The in-place `_DB_PERSISTED_MARKER` fix merged via PR #98997 (1f2bd9e763) with your Co-authored-by credit — you were the earliest carrier of this fix. #98997 supersedes this PR by centralizing the stamp into one shared helper covering all three archive_and_compact callers (micro-compact and the proactive prune had duplicate inline loops). Thank you @liuhao1024 — second fix of yours to land today (#97006 → #98543).

PR #31361, opened 2026-05-24
fix(kanban): make idempotent task creation atomic

Dusk1e:
## Summary
This change makes `create_task()` truly idempotent under concurrency by moving the `idempotency_key` lookup into the same `BEGIN IMMEDIATE` write transaction as the insert. That closes the race where two concurrent writers could both miss the pre-transaction lookup and create duplicate active tasks.

A deterministic regression test was added to force two concurrent creators through the same critical section and assert that both calls return the same `task_id` and only one row is persisted.

## Scope
- `hermes_cli/kanban_db.py`
- `tests/hermes_cli/test_kanban_db.py`

## Validation
- ✅ `uv run pytest tests/hermes_cli/test_kanban_db.py -k idempotency --timeout-method=thread`
  - Result: `1 passed, 165 deselected`
- ✅ `uv run pytest tests/hermes_cli/test_kanban_db.py -k "create_task_no_parents_is_ready or create_task_with_parent_is_todo_until_parent_done or create_task_idempotency_key_is_atomic_under_concurrency" --timeout-method=thread`
  - Result: `3 passed, 163 deselected`
- ⚠️ `uv run pytest tests/hermes_cli/test_kanban_db.py --timeout-method=thread`
  - Result: `163 passed, 3 failed`
  - Failing tests:
    - `test_resolve_hermes_argv_prefers_path_shim`
    - `test_resolve_hermes_argv_hermes_bin_bare_name_uses_path`
    - `test_resolve_hermes_argv_falls_back_to_module_form_when_no_path_shim`
  - Note: these failures are outside this patch and were triggered locally on this Windows environment after `uv run` created `.venv\Scripts\hermes.EXE`, which changes `_resolve_hermes_argv()` path resolution behavior.

## Risk
- Low.
- No schema change.
- Behavior change is isolated to the idempotent create path.

## Reviewer Note
Expected behavior after this patch: two concurrent `create_task(..., idempotency_key="same-key")` calls now converge on a single persisted task instead of producing duplicate rows.

PR #98461, opened 2026-08-30
fix(kanban): make idempotent creates atomic

MaartenDMT:
## Summary

- keep the idempotency lookup and task insert inside the same `BEGIN IMMEDIATE` transaction
- make concurrent callers return one task ID without duplicate `created` events
- add thread-level and real CLI subprocess concurrency regressions

## Why

The previous lookup ran before the write transaction. Two creators could both miss the key and insert duplicate active tasks. Serializing the lookup and insert under the existing writer transaction closes that race without a schema migration.

## Verification

- `python -m pytest tests/hermes_cli/test_kanban_cli.py tests/hermes_cli/test_kanban_core_functionality.py::test_concurrent_idempotent_creates_return_one_task -q`
- 6 passed
- cross-process CLI race repeated 5 times successfully

## Compatibility

Archived keys remain reusable. Non-idempotent creates keep the existing path. No database schema change is required.

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -3404,21 +3404,6 @@ def create_task(
             )
         skills_list = cleaned
 
-    # Idempotency check — return the existing task instead of creating a
-    # duplicate. Done BEFORE entering write_txn to keep the fast path fast
-    # and to avoid holding a write lock during the lookup. Race is
-    # acceptable: two concurrent creators with the same key might both
-    # insert, at which point both rows exist but the next lookup stabilises.
-    if idempotency_key:
-        row = conn.execute(
-            "SELECT id FROM tasks WHERE idempotency_key = ? "
-            "AND status != 'archived' "
-            "ORDER BY created_at DESC LIMIT 1",
-            (idempotency_key,),
-        ).fetchone()
-        if row:
-            return row["id"]
-
     now = int(time.time())
 
     # Resolve workspace_path from board-level default_workdir when the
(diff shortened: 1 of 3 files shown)


alt-glitch:

Duplicate of #31361: both use the same BEGIN IMMEDIATE in-transaction idempotency lookup before task insertion.

vexclawx31:
Independent synthetic-DB validation of **2bab3552e02aeff0c5ba2c5d190cca17dc4ad91d**: confirmed the non-archived idempotency lookup is inside the same IMMEDIATE transaction as insertion. No production changes applied.

Additional test-only contribution, based directly on this head: https://github.com/vexclawx31/hermes-agent/commit/87bc312b00a632975db2c45f003958be6422c9cd (no competing PR). The only adaptation from our newer-tree tests is importing connection/transaction helpers from `kanban_db`, before their module extraction.

- Three rounds of four spawned OS processes, independent native connections, synchronized admission followed by concurrent claims: one task ID/row, one created event, one claimed event, one successful claim/run, and integrity `ok` each round.
- Real lock contention for both new admission and existing-key replay exhausts the native BEGIN retry budget without INSERTs, open transactions, or SQL-dump changes; replay succeeds after release. This proves at-most-one admission, not guaranteed availability under contention.
- Delegated-child authorization denies replay on both already-writable and naturally read-only handles. Physical SQLite read-only replay is separately allowed; fresh insertion fails and rolls back. Nested rollback and archive/key lifetime are also covered.

On macOS, Python 3.13.12 / SQLite 3.50.4, using the unmodified `scripts/run_tests.sh`, `--file-retries 0 -j 2 -v -rs`:
- Added regression file + `test_kanban_db.py`, `test_kanban_decompose_db.py`, `test_kanban_write_guard.py`, `test_kanban_write_txn_busy_retry.py`: **48 passed, 4 skipped**.
- This PR's `test_kanban_core_functionality.py` + `test_kanban_cli.py`: **29 passed, 1 skipped**.

Three skips are WAL variants: native safety policy selected DELETE, so WAL was not forced or validated. The other skips are Windows-only and Linux zombie detection. Only synthetic temporary databases/homes; no guard bypasses, live DBs, full-suite claim, deployment, or hosted CI run.

PR #98462, opened 2026-08-30
fix(gateway): dedup provider-error status and final response per turn — errors no longer delivered twice on adapters without send_or_update_status

### LINKED ISSUE #72131: Provider errors delivered twice on adapters without send_or_update_status (mid-run status rewrite + identical final response)

dinershtein:
Fixes #72131.

### Problem
On adapters without `send_or_update_status`, a provider failure reaches the chat **twice** as two identical persistent messages (e.g. `⏱️ The model provider is rate-limiting requests…` ×2 for one failed turn): the mid-run status callback rewrites the raw error via `_prepare_gateway_status_message` and delivers it through the plain-`send()` fallback, then the failed turn's final response is sanitized by `_sanitize_gateway_final_response` to the byte-identical text and sent again.

### Fix
Per-turn, delivery-level dedup, built to the checklist from the review that closed #72144:

- **Recognition by membership, not re-classification.** The canonical replies from `_gateway_provider_error_reply` become named constants with a closed set (`_GATEWAY_PROVIDER_ERROR_REPLIES`). Already-rewritten text is recognized by set membership — the raw-envelope matcher does not accept e.g. the rewritten rate-limit reply ("rate-limiting requests" vs "rate limited after N retries"), which is exactly the case #72144 failed to record.
- **Per-turn state on `TurnContext`.** The status bridge records a delivered reply on `ctx._provider_error_statuses_delivered` only after a **successful** plain-send fallback. Adapters **with** `send_or_update_status` are excluded: their bubble semantics (edited in place / ephemeral / cleaned up post-run) are adapter-defined, so the gateway cannot assume a persistent copy remains and keeps delivering the final response. Concurrent turns own separate `TurnContext` instances, so nothing is process-global and nothing can cross-deduplicate chats.
- **Suppression at the outer delivery path.** `_run_agent_inner` attaches the recorded texts to the turn result at a single seam; `_handle_message_with_agent` decides right after its own sanitization pass — the path that would otherwise reconstruct the reply from the raw result, the third finding on #72144 — and blanks only the outbound delivery after transcript persistence, mirroring the intentional-silence mechanism (a delivery decision, not a transcript mutation).

### Out of scope
The failed-turn variant on `send_or_update_status` adapters (the Telegram repro in https://github.com/NousResearch/hermes-agent/issues/72131#issuecomment-5465060605, where the cleanup path is opt-in and skips failed runs) is deliberately not covered: suppressing the final response there requires knowing whether the adapter's status bubble persists, which the gateway cannot assume, and the failed-run cleanup skip is documented intentional behavior ("breadcrumbs").

### Tests
`tests/gateway/test_provider_error_double_delivery.py` (16 tests):
- every canonical reply recognized; the rewritten rate-limit reply recognized **without** the envelope classifier (the #72144 regression);
- status bridge: successful fallback send records; failed send records nothing; `send_or_update_status` adapter records nothing; ordinary status text records nothing; interleaved turns keep separate per-turn state;
- end-to-end through `_handle_message_with_agent`: identical delivered status → final delivery suppressed; transcript rows byte-identical with and without the dedup; no delivered status → reply still sent; mismatched error category → not suppressed.

Both seams were verified red: disabling the suppression flag fails the end-to-end test, disabling the recording block fails the recording and interleaving tests. Full `tests/gateway` suite green locally; `ruff check` clean on the touched files.

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -741,29 +741,79 @@ def _format_exec_approval_fallback(
         + ", ".join(choices[:-1]) + f", or {choices[-1]}."
     )
 
+_GATEWAY_PROVIDER_ERROR_AUTH_REPLY = (
+    "⚠️ Provider authentication failed. Check the configured credentials; "
+    "raw provider details are in the gateway logs."
+)
+_GATEWAY_PROVIDER_ERROR_POLICY_REPLY = (
+    "⚠️ The model provider rejected the request. I kept the raw provider "
+    "error out of chat; check gateway logs for details or try rephrasing."
+)
+_GATEWAY_PROVIDER_ERROR_RATE_LIMIT_REPLY = (
+    "⏱️ The model provider is rate-limiting requests. Please wait a moment and try again."
+)
+_GATEWAY_PROVIDER_ERROR_CONNECTION_REPLY = (
+    "⚠️ The model server is not responding — it looks like the configured "
+    "model endpoint is not running or is unreachable."
+)
+_GATEWAY_PROVIDER_ERROR_GENERIC_REPLY = (
+    "⚠️ The model provider failed after retries. I kept raw provider details "
+    "out of chat; check gateway logs for diagnostics."
+)
+
+# The closed set of user-safe replies _gateway_provider_error_reply can
+# produce. Membership in this set — not _looks_like_gateway_provider_error —
+# is how already-rewritten text must be recognized: the replies are written
+# for users, and e.g. the rate-limit reply ("rate-limiting requests") does
+# not match the raw-envelope shapes ("rate limited after N retries").
+_GATEWAY_PROVIDER_ERROR_REPLIES = frozenset({
+    _GATEWAY_PROVIDER_ERROR_AUTH_REPLY,
+    _GATEWAY_PROVIDER_ERROR_POLICY_REPLY,
+    _GATEWAY_PROVIDER_ERROR_RATE_LIMIT_REPLY,
+    _GATEWAY_PROVIDER_ERROR_CONNECTION_REPLY,
+    _GATEWAY_PROVIDER_ERROR_GENERIC_REPLY,
+})
+
+
 def _gateway_provider_error_reply(text: str) -> str:
     """Map raw provider/API errors to a short user-safe Telegram reply."""
     if _GATEWAY_AUTH_ERROR_RE.search(text):
-        return (
-            "⚠️ Provider authentication failed. Check the configured credentials; "
-            "raw provider details are in the gateway logs."
-        )
+        return _GATEWAY_PROVIDER_ERROR_AUTH_REPLY
     if _GATEWAY_PROVIDER_POLICY_RE.search(text):
-        return (
-            "⚠️ The model provider rejected the request. I kept the raw provider "
-            "error out of chat; check gateway logs for details or try rephrasing."
-        )
+        return _GATEWAY_PROVIDER_ERROR_POLICY_REPLY
     if _GATEWAY_RATE_LIMIT_RE.search(text):
-        return "⏱️ The model provider is rate-limiting requests. Please wait a moment and try again."
+        return _GATEWAY_PROVIDER_ERROR_RATE_LIMIT_REPLY
     if _GATEWAY_CONNECTION_ERROR_RE.search(text):
-        return (
-            "⚠️ The model server is not responding — it looks like the configured "
-            "model endpoint is not running or is unreachable."
-        )
-    return (
-        "⚠️ The model provider failed after retries. I kept raw provider details "
-        "out of chat; check gateway logs for diagnostics."
-    )
+        return _GATEWAY_PROVIDER_ERROR_CONNECTION_REPLY
+    return _GATEWAY_PROVIDER_ERROR_GENERIC_REPLY
+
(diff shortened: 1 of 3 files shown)


Enough1122:

**98462 — Deduplicate provider-error final delivery on fallback adapters (#72131)**

- `gateway/run.py:9-98` extracts 5 canonical user-safe replies into constants (`_GATEWAY_PROVIDER_ERROR_*`) and `frozenset _GATEWAY_PROVIDER_ERROR_REPLIES`. Adds `_is_gateway_provider_error_reply()` (set membership, not regex — rate-limit reply `"rate-limiting requests"` doesn't match raw envelope `"rate limited after N retries"`) and `_should_suppress_provider_error_final()` to detect byte-identical final already delivered as status.
- `gateway/run.py:106-131` in `TurnRunner._status_callback_sync` after `safe_schedule_threadsafe` records `_reply_text` into `ctx._provider_error_statuses_delivered` only when adapter lacks `send_or_update_status` AND `_is_gateway_provider_error_reply(prepared_message)` AND send result `success==True` (done-callback). Per-turn, thread-safe on event-loop.
- `gateway/run.py:139-158` in `_handle_message_with_agent` computes `_provider_error_status_dup` before decoration via `_should_suppress_provider_error_final(response, agent_result.get(...))`, deciding while `response` is canonical text.
- `gateway/run.py:166-179` suppresses duplicate by setting `response=""` after transcript persistence, logging `"Suppressing duplicate..."` — delivery-only, transcript untouched (mirrors `intentional_silence`).
- `gateway/run.py:187-196` plumbs `turn_ctx._provider_error_statuses_delivered` into `response["provider_error_statuses_delivered"]` for outer path.
- `gateway/turn_context.py:209-218` adds `List[str]` field `_provider_error_statuses_delivered` per-turn.
- Tests `tests/gateway/test_provider_error_double_delivery.py:227-595` thorough: canonical recognition, rate-limit regression, suppression only on identical delivered, status-bridge recording only on successful fallback send, never for `send_or_update_status` adapters, per-turn isolation, outer-path suppression with transcript untouched, mismatched category not suppressed.

Non-blocking:
- Large diff but additive and well-scattered; no breaking change to editable-bubble adapters. **Non-blocking**.
- Ensure `safe_schedule_threadsafe` monkeypatch in tests matches production future shape (`success` attr). **Non-blocking**.

Verdict: Correct fix for double-delivery of provider errors with comprehensive test matrix.

---

dinershtein:
Addressed the stub-shape note in 051f5c11c: the scheduler stubs now resolve to the real `gateway.platforms.base.SendResult`, so the tests break loudly if its shape ever drifts.

dinershtein:
@teknium1 @GottZ Could one of you approve the workflow runs on the current head (051f5c1)? All CI / Docker / Nix runs on this PR are sitting in "action required" and have never executed. The AI review's only non-blocking note (stub shape) was addressed in the last commit, and the branch is mergeable against main. Thanks!

PR #85559, opened 2026-08-13
fix: shared entry-point classification so PluginManager + provider registry agree on ownership

spfcraze:
## What & why

The `hermes_agent.plugins` entry-point group has two consumers that must agree on who owns each entry point: the general PluginManager and the model-provider registry. Two independent heuristics (manager source-kind inference, registry callable-arity check) could drift, causing:

1. **Double-load**: an enabled pip model-provider was loaded by the registry AND imported again by the general manager as a standalone plugin and flagged errored — entry-point manifests never carry `kind`, so the model-provider skip never fired.
2. **Inverse double-load**: a general plugin with a zero-arg `register()` was wrongly invoked by the provider registry as a model provider, and a pip memory-provider was invoked as a MODEL provider (wrong discovery system).

## What changed

- **New `hermes_cli/entrypoint_kind.py`** — one source-heuristic classifier (`standalone` / `exclusive` / `model-provider` / `unknown`) that reads an entry-point module's source WITHOUT importing it (a pip provider pulling in onnxruntime costs ~60MB RSS on every startup).
- **PluginManager** uses it: model-provider and memory-provider (exclusive) entry points are skipped instead of double-loaded.
- **Provider registry** uses it: general plugins and memory providers are no longer wrongly invoked as model providers. `unknown` falls through to the historical arity check so a real provider is never dropped just because its source could not be read.

## Relationship to #76567

#76567 fixes the general-manager half of this bug class (classifying entry points so the PluginManager does not import pip providers). This PR is complementary: it additionally closes the **provider-registry half** — the inverse double-load where the registry wrongly invokes general/memory plugins as model providers — and unifies both consumers on a single shared classifier (`hermes_cli/entrypoint_kind.py`) so the two sides cannot drift. Both are needed to fully close the shared-group ownership bug.

## Verification

- New regression tests cover: model-provider, memory-provider, general-plugin classification; the manager carrying kind into manifests; and the provider registry skipping a general zero-arg plugin while still loading a real model-provider.
- Full plugin + provider suites pass (482 tests), no regressions.
- No entry-point module is imported during classification.

PR #97446, opened 2026-08-28
fix(telegram): preserve durable ingress across reconnects

bgrablin:
## What does this PR do?

Telegram polling could acknowledge authorized actionable updates before the gateway had durable ownership. A process loss, reconnect, cancellation, busy deferral, or plugin-handler rebuild could therefore lose work, replay stale work, or strand a claim.

This change persists authorized actionable updates in a bot-account-scoped SQLite inbox before PTB dispatch or polling progress. It adds fenced leases, ordered replay, cancellation recovery, same-process reconnect handoff, callback quarantine, and compatibility checks for the existing SQLite schema. Plugin-installed handlers use the same durable boundary.

No configuration keys are added. The candidate changes six files: `gateway/run.py`, the Telegram adapter and inbound store, and three focused test files.

## Related Issue

No direct issue identified.

## Type of Change

- ☑ 🐛 Bug fix (non-breaking change that fixes an issue)
- ☐ ✨ New feature (non-breaking change that adds functionality)
- ☐ 🔒 Security fix
- ☐ 📝 Documentation update
- ☐ ✅ Tests only
- ☐ ♻️ Refactor (no behavior change)
- ☐ 🎯 New skill (bundled or hub)

## Changes Made

- Persist authorized actionable Telegram updates before PTB dispatch or polling progress.
- Use account-scoped SQLite ownership, fenced leases, ordered replay, retry state, quarantine, and migration preflight checks.
- Recover in-flight persistence and lease transitions across cancellation, failure, and same-process reconnect handoff.
- Route plugin-installed handlers through the durable boundary, including PTB application rebuilds.
- Add focused regressions for durable ownership, cancellation, reconnect and handoff, migration compatibility, replay, and non-Telegram queue behavior.

## How to Test

1. Run the focused suite:

   ```bash
   scripts/run_tests.sh -q tests/gateway/test_queue_consumption.py tests/gateway/test_telegram_handler_dedup.py tests/gateway/test_telegram_inbound_store.py tests/gateway/test_config.py
   ```

2. Run Ruff on the requested scope:

   ```bash
   ruff check gateway/config.py gateway/run.py plugins/platforms/telegram/adapter.py plugins/platforms/telegram/inbound_store.py tests/gateway/test_config.py tests/gateway/test_queue_consumption.py tests/gateway/test_telegram_handler_dedup.py tests/gateway/test_telegram_inbound_store.py
   ```

3. Check the diff:

   ```bash
   git diff --check origin/main...HEAD
   ```

### Code

[...]

PR #87665, opened 2026-08-16
fix(tts): let MiniMax infer emotion unless one is configured

AlexxRussell:
## What does this PR do?

`_generate_minimax_tts` always sends `voice_setting.emotion` to the t2a_v2 endpoint, falling back to `"neutral"` when the config sets nothing. MiniMax's API reference documents the opposite default: with the field omitted, the model "automatically selects the most natural emotion based on text". So the integration's fallback silently disables that inference on every request, and every MiniMax voice speaks with pinned neutral delivery regardless of what it is saying.

**Fix.** A new pure helper, `_minimax_voice_setting`, builds the `voice_setting` payload and includes `emotion` only when a config sets an explicit value. An empty value or `"auto"` also omits the field, so a config can state the intent explicitly, and explicit values are case folded before sending. The t2a_v2 payload builder calls the helper; the legacy flat `text_to_speech` branch never sent emotion and is untouched.

**This restores original behavior rather than inventing a new default.** The first MiniMax integration omitted the emotion field entirely. The hardcoded `"neutral"` arrived in `c875c0dc1` ("update MiniMax default model to speech-02 and correct API endpoint"), whose message is about the dead endpoint and model rename and never mentions emotion, so the pinning was an incidental byproduct of the endpoint correction, not a decision. MiniMax's own current request example also omits the field.

**Behavior change, disclosed.** Default output still changes for anyone who started using MiniMax after `c875c0dc1`: clips move from pinned neutral to model inferred emotion. Anyone who wants the flat sound keeps it with one line:

```yaml
tts:
  minimax:
    emotion: "neutral"
```

**Live verification.** Ran against the real API on `speech-2.8-turbo` with a cloned voice: the emotion omitted request is accepted (`base_resp.status_code 0`) and paces audibly differently from pinned neutral for identical words, confirming the field was suppressing real behavior.

## Related Issue

No existing issue. Found on a production install where a game show host voice delivered every reveal, joke and crowning in the same flat register; omitting the field restored natural delivery with no other change.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

[...]

PR #85653, opened 2026-08-13
fix(hooks): version outbound webhook envelope and fail closed on invalid config

andrexibiza:
Part of #84834 — Webhook Feature Package Task 14.

## Current disposition

The outbound-webhook lane is semantically restacked directly onto current `main` as a one-commit/three-file object.

- base: `533886c8b8eb67ff8b389b7f48e7d5e5d9c575b9`
- head: `d136e8562c372dc9cbf1dbacd94d79b2079431d3`
- commits: 1
- files: 3
- mergeable: yes

The restack was proof-preserving: comparison from the previous base to current `main` showed that none of this lane's three owned files changed upstream. The new object overlays the exact previously reviewed production/test blobs onto the exact current-main tree rather than replaying historical campaign topology.

## Owned contract

- versioned outbound webhook envelope (`schema_version: 1`);
- explicit known-field validation; inline plaintext `secret` is rejected with a value-free migration hint;
- exactly one opaque credential reference (`secret_ref`, with `secret_env` compatibility) and no raw environment fallback after the profile secret authority returns a scoped miss;
- unresolved explicit references disable the target rather than silently sending unsigned;
- V2 signs `<timestamp>.<raw body>` and receiver documentation requires freshness, constant-time HMAC verification, and delivery-ID deduplication; timestamp binding alone is not represented as replay prevention;
- redirects are refused and delivery remains bounded/fire-and-forget through the existing worker queue.

This closes the cross-profile authority defect in the old implementation: `agent.secret_scope.get_secret` is the sole credential resolver. `UnscopedSecretError` fails the target closed; unexpected runtime/import failures remain visible instead of being swallowed by a blanket `except Exception`.

## Exact-head verification

All hosted acceptance receipts are attached directly to `d136e8562c372dc9cbf1dbacd94d79b2079431d3`:

- CI `32446033866` — **success**
- Docker `32446033398` — **success**
- Nix `32446033397` — **success**

No status is inherited from the prior head. The former exact-head Windows progress failure belonged to `71c40ca...`; the current-main restack now passes that repository matrix on its own exact object.

## Topology

This remains the Task 14 outbound-hook authority under #84834. It does not absorb inbound webhook intake (#90236/#90995), verifier authority (#85318), session policy (#90304), callback transport (#85675), or final Task 19 integration (#85640).

PR #98463, opened 2026-08-30
fix(providers): skip platform and memory plugins in entry-point discovery

### LINKED ISSUE #98438: Provider discovery imports *every* enabled plugin — including platform plugins — while `gateway.config` is still loading, which breaks pip-installed platform plugins with a circular import
### What happens

I maintain a small **platform plugin** (a WhatsApp adapter for Twilio —
https://github.com/srinathh/twilio-whatsapp-platform-adapter) that is installed with
pip and registers itself through the `hermes_agent.plugins` entry-point group, exactly as the
plugin docs describe. It worked on v2026.8.3. After moving to v2026.8.27 the platform silently
disappears, and this warning shows up whenever something imports Hermes's `gateway` package
before anything else:

```
WARNING providers: Failed to load entry-point provider plugin 'twilio_whatsapp-platform':
cannot import name 'Platform' from partially initialized module 'gateway.config'
(most likely due to a circular import)
```

### Why it happens

Two pieces of Hermes both look at the same entry-point group, but they treat it differently.

1. **The plugin manager** (`hermes_cli/plugins.py`) is careful: it works out what *kind* of
   plugin each entry point is **without importing it** (`discover_entrypoint_manifests`,
   `_detect_kind_from_source`), and it loads platform plugins **lazily**, only when the
   gateway actually asks for that platform. Bundled platform plugins under
   `plugins/platforms/` rely on this — they all import `gateway.config` at the top of their
   adapter module, and that is fine because they are loaded late.

2. **The model-provider registry** (`providers/__init__.py`, added in #85504, first shipped in
   v2026.8.13) is not careful. Its `_discover_entry_point_providers()` was written so that
   pip-installed *model providers* get registered. But it calls `ep.load()` on **every**
   entry point listed in `plugins.enabled`, no matter what kind it is. There is no kind
   check before the import; the only guard (`_requires_arguments()`) runs *after* the module
   has already been imported and only applies to callables, not to bare-module entry points.

The timing makes this fatal. The registry's discovery runs from a hook that executes at
**import time** of `hermes_cli.config` (`_inject_profile_env_vars()` → `list_providers()`),
and `gateway/config.py` imports `hermes_cli.config` on its line 20 — long before it defines
the `Platform` enum (around line 325). So if the first Hermes package a process imports is
anything under `gateway.*`, the chain is:

```
import gateway.config
  └─ gateway/config.py:20      from hermes_cli.config import get_hermes_home
       └─ hermes_cli/config.py  (import-time hook) list_providers()
            └─ providers/__init__.py  _discover_entry_point_providers() → ep.load()
                 └─ my_platform_plugin/adapter.py   from gateway.config import Platform
                      → gateway.config is only half-initialised → ImportError
```

My plugin's entry module did what the documented example does — `from gateway.config import
Platform, PlatformConfig` at module top (`website/docs/developer-guide/adding-platform-adapters.md`,
around line 99) — and that is precisely what breaks. Nothing in the docs tells a plugin author
to avoid it.

### How to reproduce 

```bash
mkdir -p /tmp/stub/stubplug-0.0.1.dist-info /tmp/hh
printf 'from gateway.config import Platform\n' > /tmp/stub/stubplug.py
printf 'Metadata-Version: 2.1\nName: stubplug\nVersion: 0.0.1\n' \
    > /tmp/stub/stubplug-0.0.1.dist-info/METADATA
printf '[hermes_agent.plugins]\nstub-platform = stubplug\n' \
    > /tmp/stub/stubplug-0.0.1.dist-info/entry_points.txt
printf 'plugins:\n  enabled:\n    - stub-platform\n' > /tmp/hh/config.yaml

# Fails: gateway imported first
PYTHONPATH=/tmp/stub HERMES_HOME=/tmp/hh python -c 'import gateway.config'
#   WARNING ... Failed to load entry-point provider plugin 'stub-platform':
#   cannot import name 'Platform' from partially initialized module 'gateway.config'

# Works: hermes_cli.config imported first
PYTHONPATH=/tmp/stub HERMES_HOME=/tmp/hh python -c 'import hermes_cli.config; import stubplug'
```

Anything whose first Hermes import is under `gateway.*` hits it — test suites, tooling, and
documented commands such as `python -m gateway.platforms.feishu_comment_rules …`
(`website/docs/user-guide/messaging/feishu.md`). There is also a quieter failure mode: when the
*plugin package itself* is imported first, the registry finds the half-initialised module already
in `sys.modules` and silently registers that.

### What I would expect

Provider discovery should only import entry points that are model providers — it could reuse
the import-free classification the plugin manager already trusts, or the `<name>-platform`
naming convention the manager already relies on to derive platform ids — and/or it should not
run inside the `gateway.config` → `hermes_cli.config` import window.

### Related

- #85559 (open): "shared entry-point classification so PluginManager + provider registry agree
  on ownership" — would fix this by classifying before `ep.load()`, but it has drifted out of
  sync with `main`.
- #89199 (open): spurious `${env:VAR} not set` warnings — traced in the comments to the same
  import-time provider discovery on the config-import path. Same root, different symptom.
- The docs (`model-provider-plugin.md` "Distribute via pip", `plugins/index.md`) do not mention
  that an entry module must avoid `gateway.*` imports.

### Workaround for plugin authors

Keep the entry module free of `gateway.*` imports and import the adapter class inside
`register(ctx)`. That is what I shipped in https://github.com/srinathh/twilio-whatsapp-platform-adapter/releases/tag/v0.2.1 and it works on both v2026.8.3 and v2026.8.27.

liuhao1024:
## What does this PR do?

`_discover_entry_point_providers()` in `providers/__init__.py` calls `ep.load()` on **every** entry point listed in `plugins.enabled`, with no kind check before the import. The shared `hermes_agent.plugins` group also contains platform and memory-provider plugins, which are owned by the PluginManager / `plugins/memory` discovery and are meant to be loaded lazily. Importing a platform adapter here is actively harmful: documented adapters do `from gateway.config import Platform` at module top, and when provider discovery runs inside the `gateway.config -> hermes_cli.config` import chain (the provider-env injection at the bottom of `hermes_cli/config.py`), `gateway.config` is still half-initialized — the import fails with a circular-import error and the platform silently disappears (#98438).

This PR adds an **import-free ownership precheck** before `ep.load()` and skips entry points that are provably not model providers:

- a `<name>-platform` entry-point name — the naming convention the PluginManager itself relies on to derive platform ids (`_platform_name_from_manifest`), or
- a memory-provider source signature (`exclusive`), detected with the same import-free classifier the PluginManager already trusts (`_classify_entrypoint_value_kind`).

Anything else keeps the historical load path unchanged. This is deliberately conservative (blacklist of provable non-providers rather than a whitelist): a thin-`__init__` or lazily-importing pip provider whose source cannot be classified as `model-provider` is still loaded exactly as before, so no working provider gets deregistered. The broader "shared classifier for both consumers" refactor in #85559 addresses the same group-ownership problem from the other side; this fix is intentionally scoped to the provider-registry side and is compatible with that direction.

## Related Issue

Fixes #98438

## Type of Change

## Changes Made

- `providers/__init__.py` — in `_discover_entry_point_providers()`, classify each enabled entry point before `ep.load()`: skip `-platform`-named entry points and memory-provider (`exclusive`) source signatures; classification failures fall through to the existing load path (fail-open, ~29 lines).
- `tests/providers/test_entry_point_discovery.py` — `_FakeEP` now carries a `value` (entry-point target) like a real `EntryPoint`; added three regression tests: platform-named entry point never imported, memory-provider-signature entry point never imported, unresolvable entry point still loaded (fail-open).

## How to Test

1. Reproduce on `main` (from the issue): create a stub pip plugin named `stub-platform` whose entry module does `from gateway.config import Platform`, enable it, then `PYTHONPATH=/tmp/stub HERMES_HOME=/tmp/hh python -c "import gateway.config"` → Observed result: `WARNING providers: Failed to load entry-point provider plugin 'stub-platform': cannot import name 'Platform' from partially initialized module 'gateway.config' (most likely due to a circular import)`.
2. Same repro on this branch → Observed result: clean import, no warning, and `import hermes_cli.config; import stubplug` still works (the pre-fix working path is unchanged).
3. `pytest tests/providers/ tests/plugins/memory/test_discovery_sources.py tests/agent/test_memory_provider.py -q` → 158 passed (includes the 3 new tests; verified they fail on unpatched code for the two skip cases).

### Code

- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate — #85559 targets the same shared-group ownership with a two-sided shared classifier and is currently conflicting/stale; this PR is the minimal registry-side fix and notes the relationship above
- [x] I've run `pytest tests/ -q` and all tests pass (targeted suites: `tests/providers/`, memory-discovery, plugin-related `tests/hermes_cli/` — the batch `tests/hermes_cli/ -k plugin/provider` has 27 order-dependent failures reproducible identically on clean `main` at the same commit; each passes in isolation)
- [x] I've tested on my platform: macOS 26.4 (arm64), Python 3.11.15

- [ ] I've updated relevant documentation (README, `docs/`, docstrings) — code comments document the precheck; the plugin docs' "avoid `gateway.*` imports" caveat is a docs question for maintainers, out of scope here
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — import-free source resolution uses stdlib `importlib.util.find_spec` + `pathlib`, no platform-specific behavior

## Screenshots / Logs

Before (main) vs after (this branch) for the issue repro `PYTHONPATH=/tmp/stub HERMES_HOME=/tmp/hh python -c "import gateway.config"`:

```
# main
WARNING providers: Failed to load entry-point provider plugin 'stub-platform':
cannot import name 'Platform' from partially initialized module 'gateway.config'
(most likely due to a circular import)

# this branch
(no output — import succeeds, platform plugin left for the PluginManager's lazy load)
```

### DIFF
--- a/providers/__init__.py
+++ b/providers/__init__.py
@@ -214,6 +214,35 @@ def _discover_entry_point_providers() -> None:
                 "entry-point provider %r skipped: not enabled in config", ep.name
             )
             continue
+        # Import-free ownership precheck (#98438): this group is shared with
+        # the general PluginManager, which classifies entry points WITHOUT
+        # importing them and loads platform plugins lazily. Importing a
+        # platform adapter here breaks it — documented adapters do
+        # ``from gateway.config import Platform`` at module top, and when
+        # this scan runs inside the ``gateway.config -> hermes_cli.config``
+        # import chain (provider env injection at the bottom of
+        # hermes_cli/config.py) ``gateway.config`` is still half-initialized,
+        # so the import fails and the platform silently disappears. Skip
+        # entry points that are provably NOT model providers: a
+        # ``<name>-platform`` name (the manager's platform-id convention) or
+        # a memory-provider source signature. Anything else falls through to
+        # the load below — a real provider must never be dropped just
+        # because its source could not be classified.
+        try:
+            from hermes_cli.plugins import _classify_entrypoint_value_kind
+
+            value = getattr(ep, "value", "")
+            if ep.name.endswith("-platform") or (
+                value and _classify_entrypoint_value_kind(value) == "exclusive"
+            ):
+                logger.debug(
+                    "entry-point %r skipped by provider scan: platform or "
+                    "memory-provider plugin owned by the PluginManager",
+                    ep.name,
+                )
+                continue
+        except Exception:
+            pass  # classification unavailable — keep the historical behavior
         try:
             loaded = ep.load()
         except Exception as exc:
--- a/providers/__init__.py
+++ b/providers/__init__.py
@@ -47,6 +47,11 @@ _ALIASES: dict[str, str] = {}
 _PROVIDER_LIST_CACHE: list[ProviderProfile] | None = None
 _discovered = False
 
+# Import-free platform-adapter signal: every documented adapter does this
+# import at module top (see gateway/platforms/*), while model-provider
+# plugins never touch gateway.config. Scanned, never executed.
+_PLATFORM_IMPORT_MARKER = "from gateway.config import Platform"
+
 # Repo-root ``plugins/model-providers/`` — populated at discovery time.
 _BUNDLED_PLUGINS_DIR = (
     Path(__file__).resolve().parent.parent / "plugins" / "model-providers"
@@ -224,17 +229,31 @@ def _discover_entry_point_providers() -> None:
         # hermes_cli/config.py) ``gateway.config`` is still half-initialized,
         # so the import fails and the platform silently disappears. Skip
         # entry points that are provably NOT model providers: a
-        # ``<name>-platform`` name (the manager's platform-id convention) or
-        # a memory-provider source signature. Anything else falls through to
-        # the load below — a real provider must never be dropped just
-        # because its source could not be classified.
+        # ``<name>-platform`` name (the manager's platform-id convention), a
+        # memory-provider source signature, or the documented adapter import
+        # in the source (hand-written ``plugins.enabled`` entries may carry
+        # any name, so the suffix alone can miss them). Anything else falls
+        # through to the load below — a real provider must never be dropped
+        # just because its source could not be classified.
         try:
-            from hermes_cli.plugins import _classify_entrypoint_value_kind
+            from hermes_cli.plugins import (
+                _classify_entrypoint_value_kind,
+                _resolve_module_source,
+            )
 
             value = getattr(ep, "value", "")
-            if ep.name.endswith("-platform") or (
+            skip = ep.name.endswith("-platform") or (
                 value and _classify_entrypoint_value_kind(value) == "exclusive"
-            ):
+            )
+            if not skip and value:
+                try:
+                    module_name = str(value).split(":", 1)[0].strip()
+                    skip = bool(module_name) and _PLATFORM_IMPORT_MARKER in (
+                        _resolve_module_source(module_name)
+                    )
+                except Exception:
+                    skip = False  # unresolvable source — fail open
+            if skip:
                 logger.debug(
                     "entry-point %r skipped by provider scan: platform or "
                     "memory-provider plugin owned by the PluginManager",
--- a/tests/providers/test_entry_point_discovery.py
+++ b/tests/providers/test_entry_point_discovery.py
@@ -51,9 +51,10 @@ def _restore_real_discovery():
 
 
 class _FakeEP:
-    def __init__(self, name, loader):
+    def __init__(self, name, loader, value=""):
         self.name = name
         self.group = "hermes_agent.plugins"
+        self.value = value
         self._loader = loader
 
     def load(self):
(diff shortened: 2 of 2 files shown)


liuhao1024:
The only failing check is `OS-specific tests / Windows-only tests`, on `tests/test_desktop_update_windows_progress.py::test_progress_advances_while_the_orchestrator_blocks` — `AssertionError: /progress unresponsive until deadline (TimeoutError: timed out)`.

This looks like the known Windows-only flake rather than something this PR can influence:

- The failure is a deadline timeout in the desktop-update progress self-test; this PR only touches `providers/__init__.py` (import-free precheck in `_discover_entry_point_providers`) plus its unit tests — no code path into desktop update or its progress endpoint.
- The same test has failed on unrelated PRs (e.g. #97446 telegram, #87665 tts, #90378 gateway, #85653 hooks), and #90371 / #91255 were both merged specifically to de-flake it — it has a history of transient `/progress` stalls on CI runners.

All other checks pass on this head (Python tests, e2e, macOS-only tests, lints, builds). A maintainer rerun should clear it; happy to push any change if it fails the same way twice on an otherwise green head.

kokhlo:
Verified against the main tree: `_resolve_module_source` is import-free (reads via `resolve_module_origin`), `_detect_kind_from_source` only distinguishes `exclusive` / `model-provider`, and the `-platform` name suffix is the only import-free platform signal — so the precheck cannot accidentally drop a real provider, and the `except Exception: pass` fallback preserves the historical path when classification is unavailable. Solid, conservative fix.

One hardening thought, non-blocking: the name check `ep.name.endswith('-platform')` covers the PluginManager's naming convention, but a platform adapter registered under a name without the suffix (hand-written `plugins.enabled` entry) would still be imported. `_detect_kind_from_source` has no platform marker today, so there's no import-free way to catch that case — if it ever matters, a one-line marker scan (e.g. `from gateway.config import Platform` in the first 8192 chars) would close it without an import. Not needed for this PR's scope.

liuhao1024:
Thanks for verifying against the main tree — and for the hardening thought, which was exactly the gap I was uncomfortable leaving: a hand-written `plugins.enabled` entry bypasses the `-platform` suffix check entirely and would hit the same half-initialized `gateway.config` import chain.

I've landed it in 560dc44c3e as a third skip signal: the already-resolved module source is scanned for the documented `from gateway.config import Platform` adapter base import (first 8192 chars via `_resolve_module_source`, still import-free). Two things I checked before adding it:

- every documented adapter (both `gateway/platforms/` and `plugins/platforms/`) carries that import at module top, in all three spellings (`Platform`, `Platform, PlatformConfig`, `as _Platform`) — plain substring match covers them;
- nothing in the model-provider domain (`agent/provider_adapters/`, third-party provider plugins) imports `gateway.config`, so the marker cannot drop a real provider.

Unresolvable sources keep the fail-open load path, and a plain plugin without any marker is covered by a guard test asserting it still loads.

kokhlo:
560dc44c3e closes the gap cleanly — re-checking the three cover-spellings against the tree sold me: every adapter form is a module-top `from gateway.config import ...` statement, so the plain substring over the first 8192 chars is exact for the documented surface and stays import-free. The `except Exception: pass` fallback also still guards the pathological case (unreadable source → falls through to the load path, same as before). LGTM from my side — nothing further from the original review.

liuhao1024:
Thanks @kokhlo for the thorough re-check and for walking the adapter spellings against the tree — that's exactly the confirmation I was hoping for. Glad the fallback semantics held up too. Nothing further needed from my side; ready for maintainer review.

PR #87159, opened 2026-08-15
fix: holographic memory CJK (Chinese/Japanese/Korean) support

edward1229:
## Problem

Facts stored in the holographic memory plugin are completely unsearchable for CJK (Chinese/Japanese/Korean) users:

- `search("科技公司")` returns empty (the FTS index swallows whole sentences into single tokens)
- `probe("张伟")` returns empty (entity extraction recognizes no Chinese names)
- `reason([a, b])` returns irrelevant facts (HRR vectors behave as near-random noise for Chinese)

## Root causes (6 layers, all verified in source)

1. **FTS tokenization**: `store.py` creates FTS tables without a tokenizer → default `unicode61` merges consecutive CJK characters into one giant token (verified via `fts5vocab`)
2. **Query routing**: `retrieval.py::_sanitize_fts_query` splits on whitespace + English stopwords, no CJK handling
3. **Trigram limitation**: the trigram tokenizer (community fix for #1) still cannot match CJK runs shorter than 3 chars — Chinese given names are typically 2 chars
4. **Entity extraction**: `store.py` entity regexes are English-only (capitalized words / English quotes / "aka") → zero extraction for Chinese
5. **HRR vectors**: `holographic.py::encode_text` splits on whitespace → whole Chinese sentences hash into a single bucket
6. **Jaccard rerank**: `retrieval.py::_tokenize` whitespace splitting → CJK query/fact token intersection is always 0

### holographic.py
- add `_cjk_tokenize()`: CJK runs split into overlapping bigrams, runs ≥3 chars also keep the full run (exact phrase match); non-CJK split on whitespace
- `encode_text` now uses `_cjk_tokenize`

### retrieval.py
- `_CJK_RE` class attribute (CJK Unicode block regex)
- `_is_short_cjk_query()`: detects queries consisting only of <3-char CJK runs
- `_cjk_fts_query()`: CJK runs ≥3 chars → phrase literal (trigram-matchable); non-CJK words keep stopword+OR logic
- `_sanitize_fts_query` delegates to the CJK path first; pure-English logic unchanged
- `_get_fts_candidates` dual-arm execution: MATCH arm (English / ≥3-char CJK) + LIKE arm (<3-char CJK runs, per-run LIKE), merged & deduped in Python; LIKE arm ordered by trust_score/updated_at; each arm fails independently

## Verification (real DB: 5 facts / 19 entities, all Chinese)

| Query | Before | After |
|---|---|---|
[...]

PR #83632, opened 2026-08-11
fix(memory): make mixed CJK memories searchable

fangliquanflq:
## What does this PR do?

Holographic memory searches now find ASCII terms and three-or-more-character CJK phrases embedded in mixed-content facts. Existing stores are upgraded automatically, avoiding silent zero-result searches without discarding stored facts, and accent-insensitive Latin searches remain supported.

### Symptom

A fact containing adjacent CJK and ASCII text is stored successfully, but searches for an embedded ASCII term or a CJK phrase return no results. Whitespace-delimited English content remains searchable.

### Impact

Users of the local holographic SQLite memory provider cannot retrieve affected mixed-content memories through full-text search even though the facts remain in the database. The measured blast radius is limited to FTS lookups for mixed-content runs; two-character queries remain outside this change because SQLite trigram indexes require at least three characters.

### Bug Cause

**Trigger:** `plugins/memory/holographic/store.py:48` / creation of `facts_fts` with the default FTS5 tokenizer

**Causal chain:**
1. A user stores a fact containing adjacent CJK and ASCII text.
2. The default `unicode61` tokenizer indexes the unseparated run as a token that does not match the later exact phrase query.
3. `MemoryStore.search_facts()` receives no FTS candidate and returns an empty result set for content that exists.

**Why it is wrong:** The index tokenization does not preserve the substring-search behavior required for mixed CJK content, so valid search terms are absent from the MATCH index.

**Working sibling / contrast:** Whitespace-delimited English text creates separate `unicode61` tokens and remains searchable.

**Ruled out:** Query sanitization preserves the tested ASCII terms. A live SQLite comparison reproduced the miss with `unicode61` and matched the same stored content with `trigram`, locating the failure in index tokenization rather than ranking or query cleanup.

### Fix

Create `facts_fts` with SQLite's trigram tokenizer and retain Latin diacritic folding. On startup, detect a legacy non-trigram virtual table, rebuild it transactionally from the existing `facts` content, and leave already-upgraded stores unchanged.

## Related Issue

Closes #83593

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `plugins/memory/holographic/store.py` - use a diacritic-folding trigram FTS index and migrate legacy indexes transactionally.
- `tests/plugins/memory/test_holographic_store.py` - cover legacy-store migration, embedded ASCII and CJK retrieval, and diacritic folding.

## How to Test

[...]

PR #97050, opened 2026-08-28
fix(holographic): add trigram CJK search for fact_store (L2)

BruceAi66:
## Problem

Chinese natural-language queries never match facts in the holographic `fact_store` (L2 memory). The FTS5 table uses the default `unicode61` tokenizer, which splits CJK text into single characters, so a query like `交易哲学是什么` cannot match facts like `交易哲学: 重质不重量...`.

This is the same class of problem as L3 session search, where `messages_fts_trigram` already provides CJK substring recall.

## Change

- Add `facts_fts_trigram` FTS5 table (trigram tokenizer, content-synced via triggers, mirroring the L3 pattern).
- Route CJK queries to it via 3-char window OR-join; boost ranking by per-fact window-hit count.
- Non-CJK queries keep the existing `unicode61` path unchanged.
- Graceful degradation if the SQLite build lacks `trigram` support (`_trigram_available` probe).

## Verification

Verified locally: CJK queries like `交易哲学是什么` / `英语学习在哪` hit the correct facts; non-CJK queries behave unchanged. Existing holographic tests still pass.

PR #84633, opened 2026-08-12
fix(memory): make holographic fact search work for Chinese queries

rogerwengch:
## Problem

The holographic memory provider's `fact_store` search is completely broken for Chinese queries. The default FTS5 `unicode61` tokenizer treats a run of CJK characters as a **single token**, so a query like `飞书` never matches a fact containing `飞书网关已配置可用，WebSocket 长连接`. Every CJK `search` returns zero results — the entire deep-recall surface is dead for Chinese-language users.

## Fix

Three changes in `plugins/memory/holographic/`:

1. **trigram tokenizer** (`store.py`) — build `facts_fts` with `tokenize='trigram'` (SQLite ≥ 3.34). Trigram does substring matching and works for CJK and Latin text alike. The tokenizer choice is version-aware: on SQLite < 3.34 the table stays `unicode61` and the LIKE fallback (below) still serves CJK queries.
2. **Automatic migration** (`store.py` `_ensure_fts_table`) — a pre-existing `unicode61` `facts_fts` is detected on first init and rebuilt with trigram, re-seeding from the `facts` table. DROP+CREATE preserves the external-content triggers (bound lazily by name) and existing data. No user action needed.
3. **LIKE fallback** (`store.py` `search_facts`, `retrieval.py` `_fts_candidates`) — when FTS5 returns nothing (2-char CJK queries trigram cannot match, or a leftover unicode61 table), fall back to a `LIKE '%query%' COLLATE NOCASE` substring scan. Facts tables are small, so a full scan is cheap.

## Tests

6 new CJK retrieval tests in `tests/plugins/memory/test_holographic_retrieval.py`:
- 3+ char CJK query matches via trigram
- 2-char CJK query matches via LIKE fallback
- Mixed CJK + Latin query recalls
- ASCII retrieval not regressed by trigram
- unicode61 → trigram migration rebuilds and re-seeds an existing database
- SQLite < 3.34 keeps unicode61 and still serves CJK via the LIKE fallback

All 57 holographic tests pass.

PR #98464, opened 2026-08-30
fix(memory): make the holographic fact store CJK-capable

pengye91:
## What & why

The holographic memory provider is unusable for CJK users out of the box — verified on a production deployment with a 27k-fact Chinese corpus: Chinese queries returned **nothing** (not "fewer results" — zero). Four layered root causes:

1. **FTS5 schema**: legacy `facts_fts` used external-content mode with the default unicode61 tokenizer. A whole Chinese sentence is ONE token under unicode61, so CJK text was completely unindexed; worse, external-content tables silently return 0 rows under trigram. → rebuild as a standalone trigram table on open (existing stores migrate automatically) and repopulate from `facts`.

2. **Query path**: natural-language CJK queries ("百合花的投资结论是什么") almost never appear verbatim in documents, and FTS5's default AND semantics tank recall. → `_sanitize_fts_query` expands CJK runs into overlapping 3-char shingles, OR-joined (any shingle hit retrieves; rerank orders); latin tokens — including hyphens — are preserved even when glued to CJK without whitespace ("MiniMax的港股代码").

3. **Sub-trigram runs**: a 2-char CJK run (most Chinese words are exactly 2 chars) can NEVER match an FTS5 trigram index — not via phrase literals either. → runs <3 chars route through a bounded LIKE supplement **merged** into the candidate pool; it fires even when longer runs already matched (previously the LIKE fallback only fired on empty results, silently dropping the short-run signal).

4. **Entity extraction**: the English-only regexes found **zero** entities in Chinese text, starving `probe`/`related`/`reason` of all structural signal. → extract CJK runs (2-8 chars), CamelCase tokens, and A-share/HK ticker patterns. Verified end-to-end: entity links for a Chinese stock name went 2 → 13.

## Evidence

Measured on a 27k-fact production corpus (Chinese, real agent sessions):
- lexical-only R@5 **63%**, hybrid 79% (the semantic leg is local work, out of scope here)
- before this change: CJK retrieval returned nothing
- all 30 holographic tests green (`tests/plugins/memory/test_holographic_retrieval.py`, `test_holographic_store.py`)

## Relationship to open PRs (duplicate-check)

- #87159 covers the retrieval-side LIKE fallback (item 3) — this PR carries that concern further (bounded scan, merge-not-replace semantics) and adds shingle expansion + entity extraction.
- #83632 covers the trigram migration (item 1) — this PR additionally handles the external-content→standalone rebuild trap and repopulation.
Happy to coordinate/cede parts if maintainers prefer those PRs — the goal is CJK users getting a working store.

## Testing

- [x] All holographic unit tests pass (30/30)
- [x] Verified on a real production store (Chinese corpus, 27k facts)
- [x] Migration path tested (legacy unicode61 external-content store rebuilds in place)

### DIFF
--- a/plugins/memory/holographic/retrieval.py
+++ b/plugins/memory/holographic/retrieval.py
@@ -541,6 +541,26 @@ class FactRetriever:
             # FTS5 MATCH can fail on malformed queries — fall back to empty
             return []
 
+        # CJK short-token fallback: the trigram tokenizer cannot match
+        # queries shorter than 3 CJK characters (e.g. "芯片", "智谱").
+        # When the query contains CJK and MATCH returned nothing, retry
+        # with substring LIKE on content+tags — the same production-proven
+        # pattern the CLS news database uses for 2-char Chinese queries.
+        if not rows and self._has_cjk(query):
+            rows = self._like_fallback(conn, query, category, min_trust, limit)
+
+        # CJK supplement: a query mixing 3+ char runs with 2-char runs
+        # ("智谱 港股代码") produces a non-empty MATCH result that silently
+        # drops the 2-char signal — merge LIKE hits for those runs into the
+        # candidate pool so the rerank sees them (dedup by fact_id).
+        sub_runs = self._sub_trigram_cjk_runs(query)
+        if sub_runs and self._has_cjk(query):
+            seen = {r["fact_id"] for r in rows}
+            for extra in self._like_fallback_runs(conn, sub_runs, category, min_trust, limit):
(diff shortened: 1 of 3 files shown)


alt-glitch:

Related: #83632, #84633, and #97050 address the same holographic CJK retrieval family with different index/query designs. This PR combines standalone trigram migration, CJK query expansion and short-run fallback, plus entity extraction; reviewer choice is needed rather than treating the approaches as duplicates.

zqy1-1:
i need it

Issue #36260, opened 2026-06-01
[Bug]: Bedrock Converse drops reasoning signatures + reads wrong schema key → interleaved thinking lost / signature-replay 400 (Bedrock analog of #35975)


fesalfayed:
### Bug Description

The native AWS Bedrock provider (`agent/bedrock_adapter.py`, Converse API) mishandles Claude extended-thinking (`reasoningContent`) blocks in three ways. The net effect is the **Bedrock analog of #35975**: signed thinking is not round-tripped, so multi-turn agentic turns with interleaved thinking + tool_use either silently lose their reasoning or are rejected by Bedrock on replay.

`is_anthropic_bedrock_model()` advertises "adaptive thinking" parity (`bedrock_adapter.py:388`), so Claude 4.x on Bedrock emits the same signed, interleaved thinking blocks the native Anthropic endpoint does — but none of that machinery exists on the Bedrock path.

**Defect 1 — wrong schema key (captures nothing on real responses).**
`normalize_converse_response` reads:

```python
elif "reasoningContent" in block:
    reasoning = block["reasoningContent"]
    if isinstance(reasoning, dict):
        thinking_text = reasoning.get("text", "")   # wrong level
```

The real Converse schema (botocore `ReasoningContentBlock`) nests the text one level deeper:

```
reasoningContent = { reasoningText: { text, signature }, redactedContent: <blob> }
```

So on a genuine Bedrock response `reasoning.get("text")` is always `""` and **nothing is captured**. The fix in `8d363f8d5` (#20366 / #21202) only passes its own test because the fixture uses a flattened `{"reasoningContent": {"text": ...}}` that does not match the AWS schema.

**Defect 2 — signature dropped.**
`reasoningText.signature` is never read. AWS is explicit (botocore docs on the field): *"If you pass a reasoning block back to the API in a multi-turn conversation, include the text and its signature unmodified."* Without it, replayed thinking cannot be verified.

**Defect 3 — replay never reconstructs `reasoningContent`.**
`convert_messages_to_converse` rebuilds an assistant turn as `[text][toolUse]` only — no `reasoningContent` block is emitted, and the native path's channel (`provider_data["reasoning_details"]`, set in `agent/transports/anthropic.py`) is never populated by the Bedrock transport. So even if text + signature were captured, they would be discarded on the next turn.

[...]

PR #98469, opened 2026-08-30
fix(bedrock): stop inserting "\n\n" between every streamed reasoning delta

### LINKED ISSUE #98468: [Bug]: Bedrock streaming reasoning_content is shredded — "\n\n" inserted between every thinking delta
## Bug Description

On the Bedrock **streaming** path, the reasoning text stored on the assistant message is not what the model produced. `stream_converse_with_callbacks()` treats every `reasoningContent` **delta** as if it were a complete reasoning **block**, then joins them with the block separator `"\n\n"`. Because a delta is a fragment — routinely a fragment of a single word — the separator ends up welded between every streamed token:

```
model streamed : 'Let'  ' me list'  ...  ' divis'  'ors other than 1 and itself'
stored         : 'Let\n\n me list\n\n ...  divis\n\nors other than 1 and itself'
correct        : 'Let me list ... divisors other than 1 and itself'
```

The live display is unaffected, which is why this has gone unnoticed: `on_reasoning_delta` is handed the correct raw chunk, so the TUI renders the thinking perfectly as it streams. Only the **stored** `reasoning_content` is corrupted — and that is the copy that feeds conversation history, context compression, `/resume` replay, the API server's reasoning output, and token accounting.

This is distinct from #36260 / #89671, which are about the **non-streaming** `Converse` response reading the wrong schema key. The streaming key read here is correct; the accumulation around it is not.

### Root cause

`agent/bedrock_adapter.py` (at `2a598aad1c`). The text path immediately above buffers correctly, lines 1268-1270 and 1303-1305:

```python
            if "text" in delta:
                text = delta["text"]
                current_text_buffer.append(text)      # buffer the fragment
...
            elif current_text_buffer:
                text_parts.append("".join(current_text_buffer))   # join on block boundary
```

The reasoning path has no buffer — lines 1278-1286:

```python
            elif "reasoningContent" in delta:
                # Claude 4.6+ on Bedrock surfaces thinking via reasoningContent
                reasoning = delta["reasoningContent"]
                if isinstance(reasoning, dict):
                    thinking_text = reasoning.get("text", "")
                    if thinking_text:
                        reasoning_parts.append(str(thinking_text))   # each token = its own "part"
                        if on_reasoning_delta:
                            on_reasoning_delta(thinking_text)        # correct raw chunk
```

so at line 1327 every fragment gets a separator between it and the next:

```python
        reasoning_content="\n\n".join(reasoning_parts) if reasoning_parts else None,
```

The inflation is exactly `2 * (non_empty_deltas - 1)` bytes, which reads as merely verbose model output rather than as damage.

## Steps to Reproduce

Any Bedrock `ConverseStream` call with thinking enabled, then inspect the stored value rather than the display. Minimal, no Hermes config needed:

```python
import boto3
from agent.bedrock_adapter import stream_converse_with_callbacks

client = boto3.client("bedrock-runtime", region_name="us-east-2")
resp = client.converse_stream(
    modelId="us.anthropic.claude-sonnet-4-6",
    messages=[{"role": "user", "content": [{"text":
        "Think step by step, then answer: list the first 30 prime numbers and give their sum."}]}],
    inferenceConfig={"maxTokens": 3000, "temperature": 1.0},
    additionalModelRequestFields={"thinking": {"type": "enabled", "budget_tokens": 1500}},
)
events = list(resp["stream"])

# Ground truth: concatenating deltas in order IS the model's thinking.
deltas = [e["contentBlockDelta"]["delta"]["reasoningContent"].get("text", "")
          for e in events
          if "contentBlockDelta" in e
          and "reasoningContent" in e["contentBlockDelta"].get("delta", {})]
truth = "".join(deltas)

stored = stream_converse_with_callbacks(
    {"stream": iter(events)}).choices[0].message.reasoning_content

print(len(truth), len(stored), stored == truth)
print(repr(stored[:140]))
```

### Live run

`us-east-2`, `us.anthropic.claude-sonnet-4-6`, botocore 1.42.97, `x-amzn-RequestId: 824382e9-5ff1-4f42-8d5c-43f28914e3df`:

```
reasoningContent deltas         : 45  (44 non-empty)
distinct reasoning block indexes: [0]        <-- ONE reasoning block, not 44
ground truth  : 450 chars
stored        : 536 chars   delta +86
predicted inflation = 2 * (44 - 1) = 86      MATCH

'\n\n' occurrences   model emitted: 4    stored: 47
```

```
--- ground truth ---
'Let me list the first 30 prime numbers step by step.\n\nA prime number is a number greater than 1 that has no divisors other than 1 and itself'

--- stored by Hermes ---
'Let\n\n me list\n\n the first 30 prime numbers step\n\n by step.\n\n\n\nA prime number is a number\n\n greater than 1 that has no\n\n divis\n\nors other tha'
```

Note `divis` / `ors` — a word split across two deltas becomes `divis\n\nors`. Note also the model's own paragraph break becomes `\n\n\n\n`.

Confirming the display path is fine, so this only ever shows up in stored state:

```
on_reasoning_delta fired : 44 times
''.join(callback chunks) : 450 chars
callback stream == truth : True
```

## Expected Behavior

`reasoning_content` is the verbatim thinking text the model streamed: contiguous deltas concatenated with no inserted characters, with `"\n\n"` reserved for separating genuinely distinct reasoning blocks (distinguishable by `contentBlockIndex` / the intervening `contentBlockStop`). For the run above: 450 chars, 4 blank lines, `stored == "".join(deltas)`.

## Actual Behavior

`"\n\n"` is inserted between every non-empty delta: 536 chars, 47 blank lines, words split mid-token, and the model's own paragraph breaks doubled. Downstream, every turn's stored thinking grows by ~2 bytes per token and is no longer faithful text.

## Affected Component

Bedrock adapter — `agent/bedrock_adapter.py`, `stream_converse_with_callbacks()`. Affects every Bedrock model that streams thinking via Converse (Claude 4.6+, and the bearer-token Claude fallback that routes through Converse rather than the Anthropic SDK).

## Environment

- hermes-agent `2a598aad1c` (current `main`)
- boto3 1.42.89 / botocore 1.42.97, Python 3.11
- Amazon Linux 2023, EC2 `r7a.large`, region `us-east-2`, live Bedrock via instance role

Happy to open a PR — the fix is symmetric with the text path in the same loop (buffer, then `"".join` on the block boundary), plus a regression test built on the captured delta sequence above.

JiaDe-Wu:
Fixes #98468

## Problem

`stream_converse_with_callbacks()` treats every Bedrock `reasoningContent` **delta** as a complete reasoning **block**, appending each one to `reasoning_parts` and then joining that list with the block separator `"\n\n"`. A delta is a fragment — routinely a fragment of a single word — so the separator ends up between every streamed token:

```
model streamed : 'Let'  ' me list'  ...  ' divis'  'ors other than 1 and itself'
stored         : 'Let\n\n me list\n\n ...  divis\n\nors other than 1 and itself'
```

It has stayed invisible because `on_reasoning_delta` is handed the correct raw chunk, so the live TUI is perfect. Only the **stored** `reasoning_content` is damaged — the copy that feeds conversation history, context compression, `/resume` replay, the API server's reasoning output and token accounting.

## Fix

Mirror the text path in the same loop. `agent/bedrock_adapter.py`, +18/−2:

- add `current_reasoning_buffer` alongside `current_text_buffer`
- append deltas to the buffer instead of to `reasoning_parts`
- `"".join()` it on `contentBlockStop`, so distinct reasoning blocks still get `"\n\n"` between them and fragments within a block get nothing
- flush at the end too, so an interrupt or truncated stream still yields the partial thinking rather than dropping it

`on_reasoning_delta` still fires once per raw chunk, unchanged — that half of this path was already correct and there is a test pinning it.

Deliberately **not** touching `normalize_converse_response()`. The non-streaming path has a separate, genuinely different bug (it reads the flat `reasoning.get("text")` where the non-streaming schema nests text under `reasoningText`) which is #36260 and already has a fix in flight in #89671. Keeping this PR to the streaming accumulator so the two don't collide.

## Evidence

<details>
<summary><b>Live Bedrock run — inflation is exactly 2 × (deltas − 1)</b></summary>

`us-east-2`, `us.anthropic.claude-sonnet-4-6`, thinking enabled, botocore 1.42.97, on an EC2 instance role against live Bedrock. `x-amzn-RequestId: 824382e9-5ff1-4f42-8d5c-43f28914e3df`. Same event list replayed through both accumulators, so the only variable is the code:

```
  total events                    : 119
  reasoningContent deltas         : 45  (44 non-empty)
  distinct reasoning block indexes: [0]        <-- ONE reasoning block, not 44
  ground truth len(''.join)       : 450 chars

  ground truth   :   450 chars
  OLD (main)     :   536 chars   delta +86
  NEW (this PR)  :   450 chars   delta +0
  predicted OLD inflation = 2 * (non-empty deltas - 1) = 2 * (44 - 1) = 86
  actual   OLD inflation                               = 86   MATCH

  '\n\n' occurrences  truth=4  OLD=47  NEW=4
  NEW == ground truth : True
  OLD == ground truth : False
```

</details>

<details>
<summary><b>What the corruption looks like</b></summary>

```
--- first 140 chars, ground truth ---
'Let me list the first 30 prime numbers step by step.\n\nA prime number is a number greater than 1 that has no divisors other than 1 and itself'

--- first 140 chars, OLD (main) ---
'Let\n\n me list\n\n the first 30 prime numbers step\n\n by step.\n\n\n\nA prime number is a number\n\n greater than 1 that has no\n\n divis\n\nors other tha'

--- first 140 chars, NEW (this PR) ---
'Let me list the first 30 prime numbers step by step.\n\nA prime number is a number greater than 1 that has no divisors other than 1 and itself'
```

`divis` and `ors` arrived as two deltas and are one word. The model's own paragraph break becomes `\n\n\n\n`.

</details>

<details>
<summary><b>Why nobody noticed: the callback was never broken</b></summary>

```
  on_reasoning_delta fired      : 44 times
  ''.join(callback chunks)      : 450 chars
  callback stream == truth      : True
```

The TUI renders the correct text as it streams. The bug is only in what gets persisted.

</details>

<details>
<summary><b>Tests</b></summary>

New file `tests/agent/test_bedrock_streaming_reasoning_deltas.py`, 16 tests. The primary fixture is the real 45-delta sequence from the RequestId above, kept verbatim — including its trailing empty delta, its lone `"\n"` delta and the model's own paragraph breaks, all of which must survive untouched.

**12 of the 16 fail on `main` and pass with this change:**

```
$ git stash push -- agent/bedrock_adapter.py   # revert the fix, keep the tests
$ pytest tests/agent/test_bedrock_streaming_reasoning_deltas.py -q
12 failed, 4 passed in 0.58s

$ git stash pop
$ pytest tests/agent/test_bedrock_streaming_reasoning_deltas.py -q
16 passed in 0.52s
```

The 4 that pass either way are guards on degenerate deltas (signature-only, non-dict, no reasoning at all) — kept so a future refactor can't regress them.

Coverage: verbatim reconstruction of the real stream; the exact `2 * (n - 1)` inflation arithmetic; blank-line count preserved at 4 not 47; the `divis`/`ors` mid-word split; distinct blocks still separated by `"\n\n"`; a single block not padded; flush on a stream that ends without `contentBlockStop`; flush on interrupt mid-block; reasoning and text not contaminating each other; reasoning surviving a following `toolUse` block; empty deltas creating no spurious separators.

No existing test asserted the old behavior, so nothing had to be rewritten to accommodate this.

**Full suites, this branch:**

```
$ pytest tests/agent/test_bedrock_1m_context.py tests/agent/test_bedrock_adapter.py \
         tests/agent/test_bedrock_empty_text_blocks.py tests/agent/test_bedrock_integration.py \
         tests/agent/test_bedrock_interrupt_post_worker.py \
         tests/agent/test_bedrock_streaming_reasoning_deltas.py -q
134 passed in 3.94s

$ ruff check agent/bedrock_adapter.py tests/agent/test_bedrock_streaming_reasoning_deltas.py
All checks passed!
```

`tests/run_agent/` has 7 failures on this branch. The identical 7 fail on unpatched `main` — Anthropic SDK streaming, switch-model reasoning-override, Nous fallback activation, Anthropic interrupt handler — so none are caused by this change. Verified by running the suite in two separate checkouts, `2a598aad1c` clean and this branch, and diffing the `FAILED` sets:

```
baseline (2a598aad1c) FAILED count: 7
this branch           FAILED count: 7
diff: IDENTICAL
```

</details>

## How to Test

```bash
pytest tests/agent/test_bedrock_streaming_reasoning_deltas.py -q
```

Against live Bedrock, with credentials and a thinking-capable model — the reproduction in the issue prints ground truth vs stored and asserts equality.

## Notes for review

- The buffer is flushed on `contentBlockStop` rather than only at the end, so a model that emits multiple thinking blocks in one turn (interleaved thinking) keeps them separated by `"\n\n"` exactly as the current code intends. `"\n\n"` was always the right separator — it was just being applied at the wrong granularity.
- Happy to fold in the non-streaming `reasoningText` read as well if you'd rather see #36260 closed in one PR instead of alongside #89671; I kept them apart so this one stays reviewable and doesn't conflict.

### DIFF
--- a/agent/bedrock_adapter.py
+++ b/agent/bedrock_adapter.py
@@ -1229,6 +1229,7 @@ def stream_converse_with_callbacks(
     tool_calls: List[SimpleNamespace] = []
     current_tool: Optional[Dict] = None
     current_text_buffer: List[str] = []
+    current_reasoning_buffer: List[str] = []
     has_tool_use = False
     stop_reason = "end_turn"
     usage_data: Dict[str, int] = {}
@@ -1276,16 +1277,26 @@ def stream_converse_with_callbacks(
                 if current_tool is not None:
                     current_tool["input_json"] += delta["toolUse"].get("input", "")
             elif "reasoningContent" in delta:
-                # Claude 4.6+ on Bedrock surfaces thinking via reasoningContent
+                # Claude 4.6+ on Bedrock surfaces thinking via reasoningContent.
+                # A reasoning delta carries a fragment of a word, not a whole
+                # thought — buffer it and join on the block boundary, exactly as
+                # the text path does above. Appending straight to reasoning_parts
+                # would weld the "\n\n" block separator between every token.
                 reasoning = delta["reasoningContent"]
                 if isinstance(reasoning, dict):
                     thinking_text = reasoning.get("text", "")
                     if thinking_text:
-                        reasoning_parts.append(str(thinking_text))
+                        current_reasoning_buffer.append(str(thinking_text))
                         if on_reasoning_delta:
                             on_reasoning_delta(thinking_text)
 
         elif "contentBlockStop" in event:
+            # A reasoning block just ended (or a text/tool block did, in which
+            # case the buffer is empty). Genuinely separate thinking blocks stay
+            # separated by the "\n\n" join below.
+            if current_reasoning_buffer:
+                reasoning_parts.append("".join(current_reasoning_buffer))
+                current_reasoning_buffer = []
             if current_tool is not None:
                 try:
                     input_dict = json.loads(current_tool["input_json"]) if current_tool["input_json"] else {}
@@ -1319,6 +1330,11 @@ def stream_converse_with_callbacks(
     # Flush remaining text
     if current_text_buffer:
         text_parts.append("".join(current_text_buffer))
+
+    # Flush reasoning too — an interrupt or a truncated stream can end without
+    # the closing contentBlockStop, and partial thinking still beats none.
+    if current_reasoning_buffer:
+        reasoning_parts.append("".join(current_reasoning_buffer))
 
     msg = SimpleNamespace(
         role="assistant",
--- a/tests/agent/test_bedrock_streaming_reasoning_deltas.py
+++ b/tests/agent/test_bedrock_streaming_reasoning_deltas.py
@@ -1,0 +1,216 @@
+"""Regression tests for corrupted ``reasoning_content`` on the Bedrock streaming path.
+
+``ConverseStream`` delivers thinking as a long run of ``reasoningContent`` deltas
+that each carry a *fragment* — frequently a fragment of a single word. The
+accumulator appended every fragment to ``reasoning_parts`` as though it were a
+whole reasoning block, then joined the list with ``"\\n\\n"``, so the block
+separator was welded between every streamed token::
+
+    deltas  : 'Let'  ' me list'  ' divis'  'ors other than 1 and itself'
+    stored  : 'Let\\n\\n me list\\n\\n divis\\n\\nors other than 1 and itself'
+    correct : 'Let me list divisors other than 1 and itself'
+
+Two properties made this hard to spot:
+
+* ``on_reasoning_delta`` was always handed the correct raw chunk, so the live TUI
+  rendered the thinking perfectly. Only the *stored* value was corrupted — which
+  is what feeds conversation history, context compression, replay and token
+  accounting.
+* The inflation is exactly ``2 * (non_empty_deltas - 1)`` bytes, which merely
+  looks like verbose model output rather than a bug.
+
+The fix mirrors the text path directly above it in the same loop: buffer
+contiguous deltas, ``"".join`` them on the ``contentBlockStop`` boundary, and keep
+``"\\n\\n"`` for separating genuinely distinct reasoning blocks.
+
+The primary fixture below is a real delta sequence captured from
+``us.anthropic.claude-sonnet-4-6`` in ``us-east-2``
+(``x-amzn-RequestId: 824382e9-5ff1-4f42-8d5c-43f28914e3df``, botocore 1.42.97),
+kept verbatim — including the trailing empty delta and the model's own ``\\n\\n``
+paragraph breaks, which must survive untouched.
+"""
+
+from agent.bedrock_adapter import stream_converse_with_callbacks
+
+# Captured verbatim from a live ConverseStream call with thinking enabled.
+REAL_DELTAS = [
+    "Let", " me list", " the first 30 prime numbers step", " by step.",
+    "\n\nA prime number is a number", " greater than 1 that has no", " divis",
+    "ors other than 1 and itself", ".\n\n1", ". 2", "\n2. 3\n3. ", "5\n4. 7\n5",
+    ". 11\n6. 13", "\n7. 17\n8.", " 19\n9. 23", "\n10. 29\n11. ",
+    "31\n12. 37\n13", ". 41\n14. 43", "\n15. 47\n16.", " 53\n17. 59",
+    "\n18. 61\n19. ", "67\n20. 71\n21", ". 73\n22. 79", "\n", "23. 83\n24. ",
+    "89\n25. 97\n26", ". 101\n27. 103", "\n28. 107\n29.", " 109\n30. 113",
+    "\n\nNow", " I'm", " adding", " them", " all", " together", " to",
+    " get the sum", " of the first 30 pr", "imes.", "\n\nContinuing", " the",
+    " running", " total", "...", "",
+]
+
+# Concatenating the deltas in order IS the model's thinking text, by definition
+# of a delta stream. Everything else is measured against this.
+REAL_TRUTH = "".join(REAL_DELTAS)
+
+
+def _reasoning_delta(text, index=0):
+    return {"contentBlockDelta": {"contentBlockIndex": index,
+                                  "delta": {"reasoningContent": {"text": text}}}}
+
+
+def _text_delta(text, index=0):
+    return {"contentBlockDelta": {"contentBlockIndex": index,
+                                  "delta": {"text": text}}}
+
+
+def _stop(index=0):
+    return {"contentBlockStop": {"contentBlockIndex": index}}
+
+
+def _run(events, **callbacks):
+    return stream_converse_with_callbacks({"stream": iter(events)}, **callbacks)
+
+
+def _reasoning_of(events, **callbacks):
+    return _run(events, **callbacks).choices[0].message.reasoning_content
+
+
+def _real_stream():
+    return [_reasoning_delta(d) for d in REAL_DELTAS] + [_stop()]
+
+
+# ── the real captured stream ──────────────────────────────────────────────────
+
+def test_real_stream_reasoning_is_verbatim():
+    """The stored thinking text is byte-identical to what the model streamed."""
+    assert _reasoning_of(_real_stream()) == REAL_TRUTH
+
+
+def test_real_stream_has_no_separator_inflation():
+    """No bytes are invented. The old accumulator added 2 * (44 - 1) == 86."""
+    non_empty = len([d for d in REAL_DELTAS if d])
+    assert non_empty == 44
+    assert len(_reasoning_of(_real_stream())) == len(REAL_TRUTH) == 450
+    # What the bug produced, stated explicitly so a regression is unmistakable.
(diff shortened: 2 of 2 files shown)


Enough1122:

**98469 — Buffer Bedrock `reasoningContent` deltas instead of joining each token**

- `agent/bedrock_adapter.py:9,32-50` adds `current_reasoning_buffer: List[str]` and changes `reasoningContent` handling: previously `reasoning_parts.append(str(thinking_text))` on every delta then `"

".join(...)` welded separator between every token (`'divis' + 'ors' → 'divis\n\nors'`). Now appends to `current_reasoning_buffer`, joins on `contentBlockStop` boundary (`reasoning_parts.append("".join(current_reasoning_buffer))`), mirroring text path. Also flushes buffer at loop end for truncated/interrupted streams (`if current_reasoning_buffer: reasoning_parts.append(...)`), so partial thinking not lost.
- `contentBlockStop` handling extended to drain reasoning buffer before tool handling — keeps genuine distinct blocks separated by `"

"`, fragments within block concatenated.
- Tests `tests/agent/test_bedrock_streaming_reasoning_deltas.py:54-275` use verbatim 44-delta capture from `us.anthropic.claude-sonnet-4-6` (`REAL_DELTAS`, `REAL_TRUTH` 450 chars, 4 `\n\n`). Covers: `test_real_stream_reasoning_is_verbatim`, no separator inflation (old 450+86), preserves only model's 4 paragraph breaks (bug reported 47), word split `'divis'+'ors'` not broken, leading whitespace preserved, callback still fires per raw chunk (44), distinct blocks remain `\n\n`, truncated/interrupt flush, reasoning+text independent, signature-only delta yields `None`.

Non-blocking:
- Live `on_reasoning_delta` still fires per fragment — not batched, so TUI rendering unchanged. **Non-blocking**.
- Empty deltas filtered (`if thinking_text:`) so spurious separators avoided — correct. **Non-blocking**.

Verdict: Root-causes stored reasoning corruption while live view was correct; fix mirrors text path with strong real-data tests.

---

JiaDe-Wu:
Closing this in favour of #100875, which is also mine — I duplicated my own work and did not notice until the triage bot flagged it.

What happened, plainly: I opened this on 2026-08-30. On 2026-09-02 I went looking for unclaimed issues, found #98468 with no comments, and checked the issue for signs anyone was on it — but not for a PR already linked to it. So I implemented the same fix twice. The search I told a reviewer on #98564 I should have run is the same one I skipped here, which is worth recording rather than quietly deleting a branch.

**#100875 is the one to take, and not only because it is newer.**

`main` gained a `stream_blocks` accumulator between these two PRs. That record already builds each content block's reasoning text correctly, by plain concatenation, and it backs `bedrock_content_blocks`. So on current `main` the second accumulator (`reasoning_parts`) is redundant as well as wrong, and #100875 deletes it and derives `reasoning_content` from those ordered blocks — one place decides how the stored text is built. This PR instead mirrors the text path's buffering inside the loop, which was the right shape against the tree as it stood on 30 August but adds a third accumulator to the tree as it stands now.

That difference is also why this one is `CONFLICTING/DIRTY` and #100875 is not: the code it edits has moved.

Everything else carried across. #100875 has the same measured evidence (45 real reasoning deltas from `us.anthropic.claude-sonnet-4-5-20250929-v1:0`, 88 characters the model never emitted, `"\n\n"` count 2 emitted vs 46 stored), plus a guard test pinning that `normalize_converse_response` keeps `"\n\n".join` because it walks genuine blocks, and mutation coverage on the new logic. Its tests live in the existing `tests/agent/test_bedrock_adapter.py` rather than a new file, which is closer to how the rest of that suite is organised.

Nothing is lost by closing this. If a maintainer would rather keep the earlier PR number, say so and I will rebase this one onto current `main` and close #100875 instead — the fix is the same either way and I have no attachment to which number carries it.

PR #98471, opened 2026-08-30
fix(update): persist dashboard update receipt across restart

KythornAi:
## Summary

- persist an initial `running` update receipt before any restart boundary
- recover a Dashboard-owned updater killed by the Dashboard systemd cgroup as an honest `interrupted` receipt on replacement Dashboard startup
- retain the same timestamped receipt rather than creating conflicting records

## Root cause

The Dashboard starts `hermes update` as a child in its systemd control group. Restarting `hermes-dashboard.service` kills that child before its in-memory command-boundary receipt finaliser can write `update_receipts/latest.json`.

## Validation

- `26 passed, 175 deselected`
- Targeted receipt lifecycle and Dashboard-startup recovery tests passed.
- Dashboard restart, listener and authenticated HTTP redirect were verified locally.
- No active Hermes Doctor security advisory.

### DIFF
--- a/hermes_cli/update_receipt.py
+++ b/hermes_cli/update_receipt.py
@@ -57,6 +57,7 @@ class UpdateReceipt:
     """Collects the observable facts of one ``hermes update`` run."""
 
     def __init__(self) -> None:
+        self.path: Optional[Path] = None
         self.data: dict[str, Any] = {
             "schema": 1,
             "started_at": _utc_now_iso(),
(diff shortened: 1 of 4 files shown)


Enough1122:

**98471 — Persist update receipt early & recover interrupted updater**

- `hermes_cli/update_receipt.py:9,17-91` adds `UpdateReceipt.path: Optional[Path]`, extracts `_persist_receipt(receipt)` that writes timestamped `update_<stamp>_<pid>.json` + `latest.json` pointer and prunes old receipts, called from `begin_update_receipt` (now persists `running` record before restart boundary). Adds `_receipt_pid_is_live`, `_find_persisted_receipt`, `recover_interrupted_update_receipt()` which checks `_current is None`, `latest outcome=="running"`, `pid` dead, then loads persisted file, `finalize("interrupted")`, sets `stop_reason="updater exited before finalization"`, and re-persists.
- `finalize_update_receipt:124` now delegates to `_persist_receipt`.
- `hermes_cli/web_server.py:136-156` adds `_recover_interrupted_update_receipt_at_startup()` called in `_lifespan` before `state.db` migration — handles Dashboard-owned updater child killed by service restart (systemd cgroup) before finalizer runs.
- Tests `tests/hermes_cli/test_update_receipt.py:169-197` verify `begin_persists_running_receipt_for_crash_recovery` and `recover_dead_running_receipt_as_interrupted` (simulate death via `_current=None` + `_receipt_pid_is_live→False`). `test_web_server.py:209-222` checks startup recovery wiring.

Non-blocking:
- `_persist_receipt` called inside `begin_update_receipt` try/except — failure is defensive (`return None`, debug log) and doesn't abort update. Good. **Non-blocking**.
- `_receipt_pid_is_live` imports `gateway.status._pid_exists` lazily — ensure `gateway` optional dep doesn't break `hermes update` in minimal installs (fallback returns False). **Non-blocking**.
- New globals (`receipt.path`, `_current`) are process-local; per-PID stamp avoids collision. **Non-blocking**.

Verdict: Fixes lost `running` receipt when Dashboard restart kills its own updater.

---

PR #96204, opened 2026-08-27
fix(gateway): stop hung sessions from auto-resuming on every restart

fangliquanflq:
## What does this PR do?

Prevents a restart-interrupted gateway session from being auto-resumed forever after repeated hard kills. The gateway now counts each synthetic startup resume before dispatch, so attempts survive SIGKILL and the existing stuck-session suspension threshold can retire a repeatedly hung session while the gateway continues serving new inbound messages.

### Symptom

A `resume_pending=restart_interrupted` session can consume one CPU thread until the agent idle watchdog fires, then be replayed again on every later gateway restart.

### Impact

The affected session repeatedly consumes CPU for up to the watchdog duration and cannot recover without manual session-state surgery. The startup restore gate is already bounded for other channels, but it does not prevent the same bad session from being replayed on the next boot.

### Bug Cause

**Trigger:** `gateway/run.py` / `GatewayRunner._schedule_resume_pending_sessions()` schedules a fresh synthetic turn for every eligible restart-interrupted session.

**Causal chain:**
1. An unclean systemd stop leaves a durable `resume_pending` session.
2. Startup dispatches the synthetic resume, but a SIGKILL during the hung turn bypasses the shutdown-only `.restart_failure_counts` update.
3. A later boot sees no per-session failure count and dispatches the same turn again.

**Why it is wrong:** the global restart-loop guard only chains boots whose gaps stay within its configured maximum. A resume that hangs for about an hour exceeds the default 300-second gap, so each boot begins a new global chain while the per-session shutdown counter remains unchanged.

**Working sibling / contrast:** fast crash loops are stopped by the global restart-loop guard, and graceful shutdowns update the existing per-session counter. The missing case is a long startup resume terminated by SIGKILL.

**Ruled out:** the startup restore drain bound is not the missing guard. It releases queued inbound messages after its timeout, but intentionally leaves the slow resume turn running and does not persist a failed attempt for the next boot.

### Fix

- Persist each authorized, adapter-ready startup resume attempt before creating its task, and refuse dispatch unless the increment is durably published.
- Preserve malformed or unreadable retry evidence instead of resetting it, and serialize ledger transactions so concurrent completion cannot erase a newer admission.
- Reuse the existing `.restart_failure_counts` file and successful-turn clear path rather than adding a second session counter.
- Preserve unresolved per-session counts when unrelated sessions shut down or another stuck session is suspended.
- Keep retry evidence until both stuck-session suspension and successful recovery state are durably persisted.
[...]

PR #91058, opened 2026-08-20
fix(gateway): preserve startup resume provenance

evgyur:
## Summary
- mark only the gateway-generated startup recovery event with explicit `startup_resume` provenance
- carry that provenance through the agent turn instead of inferring it from empty text or the broad `internal` flag
- keep captionless media and other empty real events on the new-user-message path while preserving native multimodal persistence

## Problem
The resume-pending path treated an empty message as the synthetic startup wake. Real inbound events can also have empty text, notably media without a caption. `internal` is not a safe substitute because it also covers plugin injections, handoffs, and background completions.

## Testing
- `HERMES_PYTHON=/opt/hermes-agent/venv/bin/python scripts/run_tests.sh tests/gateway/test_restart_resume_pending.py tests/gateway/test_run_progress_interrupt.py tests/gateway/test_matrix_message_event_metadata.py -q`
- 43 passed
- `python -m py_compile` on all changed Python files

## Scope
This keeps the existing interactive/non-interactive recovery policy. It only gives that policy exact event provenance and adds regression coverage for captionless input.

PR #71376, opened 2026-07-25
fix: offload context engine startup from gateway loop

Tosko4:
## What does this PR do?

Moves temporary `AIAgent` construction off gateway/aiohttp event loops for three direct async entry paths:

- automatic session-hygiene compression;
- manual `/compress`;
- API `POST /v1/runs`.

Context-engine `on_session_start()` hooks remain synchronous-compatible, but plugin-defined startup work now runs behind a context-preserving executor boundary. Cancellation-safe acquisition retains an in-flight constructor and cleans up the eventual agent instead of orphaning it.

This addresses the `/v1/runs` sibling path identified in [Teknium's review](https://github.com/NousResearch/hermes-agent/pull/71376#issuecomment-5129970625).

Companion plugin-side root-cause fix: [stephenschoettler/hermes-lcm#440](https://github.com/stephenschoettler/hermes-lcm/pull/440)

## Related Issue

No existing issue or PR matched the reproduced lifecycle-blocking path.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `gateway/run.py`
  - adds one context-preserving, cancellation-safe temporary-agent construction boundary;
  - retains shielded in-flight construction after caller cancellation;
  - performs the existing bounded agent cleanup off-loop once construction finishes;
  - uses the boundary for session-hygiene compression.
- `gateway/slash_commands.py`
  - constructs the temporary `/compress` agent through the same boundary.
- `gateway/platforms/api_server.py`
  - constructs `/v1/runs` agents off the aiohttp loop;
  - reuses the `GatewayRunner` boundary in normal gateway operation;
  - provides the same context-preserving/cancellation-safe fallback for standalone adapters used by tests and embedders.
- Regression tests prove:
  - blocked startup does not stall gateway or API event-loop callbacks;
  - `/compress` construction runs off-loop;
  - cancellation cannot orphan an agent that finishes constructing later;
  - `/v1/runs` uses the shared production boundary;
  - existing run lifecycle, stop, SSE, and approval-isolation behavior remains intact.

## How to Test

Run the affected gateway/context-engine surface:

```bash
scripts/run_tests.sh -j 4 \
[...]

PR #98473, opened 2026-08-30
fix(gateway): keep inbound ready during startup recovery

jaebarista:
## What does this PR do?

Keeps gateway cold-start availability independent from restart-interrupted session recovery. Platform adapters can connect and serve inbound messages while per-session auto-resume work continues in the background.

## Problem

The gateway initialized `_startup_restore_in_progress = True` before connecting adapters. That global gate queued every non-internal inbound event until startup resume tasks drained. Although the drain has a timeout, a huge transcript, context-compression failure, or stalled auto-resume could still make a live gateway appear unavailable after restart even though its process and platform adapter were alive.

The affected recovery sessions are already synchronously pre-claimed in `_schedule_resume_pending_sessions()`. That per-session slot prevents duplicate work without requiring a gateway-wide inbound gate.

## Fix

- Initialize startup restore with the global inbound gate open.
- Preserve the existing per-session pre-claim and background auto-resume lifecycle.
- Keep the restore queue/task containers initialized for compatibility with the existing bounded drain helpers.
- Add a regression covering giant-session load failure, context-compression failure, and auto-resume failure.

## Why this is safe

- Same-session work remains serialized by `_AGENT_PENDING_SENTINEL` pre-claiming.
- Other sessions no longer wait behind unrelated recovery work.
- No configuration, persistence schema, platform-specific behavior, or public API changes.
- Existing bounded startup-restore and delivery tests remain green.

## Duplicate / related work checked

Searched open and merged PRs/issues for gateway startup recovery, inbound readiness, Slack restart auto-resume, and context-compression startup. Related PRs such as #96204 (cross-boot retry budgeting), #91058 (startup-resume provenance), and #71376 (context-engine construction offload) address different failure mechanisms.

## Testing

Canonical focused suite:

```bash
scripts/run_tests.sh tests/gateway/test_restart_resume_pending.py -q
```

Result: **39 passed, 0 failed**.

Regression sabotage check: the new three-case test fails on unmodified `origin/main` and passes with this patch.

Additional local acceptance on macOS verified that a restart successor reached platform connection and inbound readiness while continuation recovery remained independently tracked.

## Checklist

- [x] Read `CONTRIBUTING.md`
- [x] Searched PRs/issues for duplicates
- [x] Conventional commit message
- [x] Focused canonical test suite passed
- [x] Regression test demonstrated failure without the fix
- [x] No config or schema changes
- [x] Cross-platform: Python/asyncio behavior only; no OS-specific core code

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -12154,6 +12154,12 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
         except Exception:
             pass
 
+    def _initialize_startup_restore_state(self) -> None:
+        """Keep cold-start recovery session-local instead of gating all inbound."""
+        self._startup_restore_in_progress = False
+        self._startup_restore_queue = []
+        self._startup_restore_tasks = []
+
     async def _drain_startup_restore_queue(self) -> int:
         """Replay inbound messages queued while startup auto-resume ran."""
         drained = 0
@@ -13308,14 +13314,14 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
         except Exception as e:
             logger.debug("Stuck-loop detection failed: %s", e)
 
-        # Serialize startup restore against inbound dispatch.  Platform
-        # adapters can begin receiving messages as soon as they connect, but
-        # restart-interrupted sessions are not auto-resumed until all startup
-        # wiring below completes.  Queue inbound messages until the resume
-        # pass runs and every synthetic resume turn has finished.
-        self._startup_restore_in_progress = True
-        self._startup_restore_queue = []
-        self._startup_restore_tasks = []
+        # Cold boot availability outranks session recovery. Do NOT globally
+        # gate inbound while restart-interrupted sessions auto-resume: each
+        # resume session is synchronously pre-claimed, so same-session inbound
+        # still queues safely behind its own slot while every other Slack
+        # conversation remains serviceable even if a huge transcript,
+        # compression failure, or model stall makes recovery pathological.
+        # Auto-resume remains a post-connect background task.
+        self._initialize_startup_restore_state()
 
         connected_count = 0
         enabled_platform_count = 0
--- a/tests/gateway/test_restart_resume_pending.py
+++ b/tests/gateway/test_restart_resume_pending.py
@@ -705,6 +705,55 @@ async def test_reconnect_reschedule_is_platform_scoped():
 
 
 @pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "failure",
+    [
+        MemoryError("giant-session-load-failed"),
+        RuntimeError("context-compression-failed"),
+        RuntimeError("auto-resume-failed"),
+    ],
+    ids=["giant-session", "compression-failure", "auto-resume-failure"],
+)
+async def test_startup_recovery_failure_never_enables_global_inbound_gate(failure):
+    """Broken recovery work is background-only after platform availability.
+
+    A huge transcript, compressor failure, or resume-agent failure may leave
+    that one session pending, but must not switch the gateway back into the
+    global startup queue that blocks Slack inbound for every session.
+    """
+    runner, adapter = make_restart_runner()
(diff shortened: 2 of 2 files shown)


Enough1122:

**98473 — Make startup restore non-blocking for cold boot**

- `gateway/run.py:9-37` extracts `_initialize_startup_restore_state()` setting `_startup_restore_in_progress=False`, `queue=[]`, `tasks=[]` (previously `True`/`[]`/`[]` inline in `start()`). Comment: cold boot availability outranks session recovery — do NOT globally gate inbound while restart-interrupted sessions auto-resume; each resume session is synchronously pre-claimed so same-session inbound still queues safely behind its slot while every other Slack conversation remains serviceable even if huge transcript/compression failure/model stall makes recovery pathological. Auto-resume stays post-connect background task.
- Test `tests/gateway/test_restart_resume_pending.py:49-96` parametrized over `MemoryError("giant-session-load-failed")`, `RuntimeError("context-compression-failed")`, `RuntimeError("auto-resume-failed")` asserts after `_initialize_startup_restore_state()` flag is `False` and after scheduling 1 pending resume (with `handle_message` raising) ` _startup_restore_in_progress` stays `False` and `pending_entry` not in `_running_agents`. Also existing `test_startup_restore_waits_for_resume_before_draining_inbound` preserved.

Non-blocking:
- Narrow diff — only changes initialization; actual queue/drain logic unchanged. Correct that per-session pre-claim still orders same-session inbound. **Non-blocking**.
- Ensure `start()` still sets up queue before `connected_count` loop; `initialize` called at same site so ordering preserved. **Non-blocking**.

Verdict: Improves availability during pathological recovery with minimal code change.

---

teknium1:
Triage note on the interaction with issue #99373 (P1, filed 2026-08-31) — concrete evidence, not a taste call:

This PR initializes `_startup_restore_in_progress = False`, removing the global inbound gate during startup. The availability premise is real (a pathological recovery turn shouldn't block every channel), but **main already bounds that exact failure**: `_finish_startup_restore` waits on boot resume turns with a 30s cap (`agent.gateway_startup_restore_drain_timeout`, non-cancelling `asyncio.wait`), releasing the gate and letting slow turns finish in the background. So the pathological-recovery scenario in this PR's description can hold the gate for at most the configured bound, not indefinitely.

Meanwhile #99373 shows the gate window is the only thing standing between an inbound message and a **skeleton system prompt** at boot (`in=1737` tokens vs ~14.6K — no context tier, no tool schemas, `check_fn` probes failing cold in-turn). I live-reproduced this on 26f178e5fa: on a fresh boot the gate opens ~0ms after being reached with `run_agent`/`model_tools` unimported and the check_fn cache empty. Removing the gate entirely would widen that window to the whole boot for every session, including queued restore replays.

PR #99516 keeps the gate, adds a **bounded turn-machinery warm-up** (default 20s, opt-out via config) before the gate opens, and preserves this PR's availability guarantee: a wedged init releases the gate at the bound and finishes warming in the background. Suggest evaluating that direction against this one before merging here — as written, this PR would make #99373 strictly worse. Not closing anything; maintainer's call.

kshitijk4poor:
Closing as **not planned** — the premise was overtaken by #99373's fix, which moved in the opposite direction.

This PR initializes `_startup_restore_in_progress = False`, removing the global inbound gate during startup recovery. Since it was opened, issue #99373 (skeleton system prompt at boot) was root-caused to inbound turns racing the turn-machinery warm-up, and the fix that merged (`fc2421cfa7`, now on main) **depends on that gate**: `_finish_startup_restore` holds inbound until a bounded warm-up (`gateway_startup_warmup_timeout`, default 20s) completes. Merging this PR would reintroduce the #99373 regression.

The pathological-recovery concern this PR targets is separately bounded on main: `_finish_startup_restore` waits on boot-resume turns with a 30s cap (`gateway_startup_restore_drain_timeout`, non-cancelling), so a stuck recovery turn holds the gate for at most the configured bound — not indefinitely.

Thanks @jaebarista for the report and the test coverage — the underlying worry (startup blocking all inbound) is real, but the bounded-gate design that landed addresses it without opening the warm-up race.

PR #98475, opened 2026-08-30
fix(desktop): stop re-persisting an exhausted session after cleanup

### LINKED ISSUE #98467: Desktop: exhausted session is re-remembered after cleanup, so every relaunch reopens the dead chat
## Summary

`useDesktopIntegrations` re-persists a session whose bounded resume retry has
already **exhausted**, immediately after the cleanup effect cleared it. The two
effects fight each other, the dead id wins, and it survives every restart — the
window boots into the resume-error screen on each cold start until the user
manually hits New chat.

This is the mechanism that makes #60541's closing assumption false:

> The pinned id is cleared (so the *next* launch lands on new-chat), but this
> launch is stranded.

It is not cleared in practice. It is written straight back.

## Root cause

`apps/desktop/src/app/contrib/hooks/use-desktop-integrations.ts` — two effects
in the same hook disagree about the same two storage keys.

**Effect 1 (cleanup)** drops the remembered navigation when the routed session's
resume gave up:

```ts
useEffect(() => {
  if (!profileReady || !resumeExhaustedSessionId) return

  if (getRememberedSessionId(activeProfile) === resumeExhaustedSessionId) {
    setRememberedSessionId(null, activeProfile)
  }
  if (routeSessionId(getRememberedRoute(activeProfile) ?? '') === resumeExhaustedSessionId) {
    setRememberedRoute(null, activeProfile)
  }
}, [activeProfile, profileReady, resumeExhaustedSessionId])
```

**Effect 2 (persistence)** has no idea the session is stranded:

```ts
if (routedSessionId && sessionBelongsToProfile(sessions, routedSessionId, activeProfile)) {
  setRememberedSessionId(routedSessionId, activeProfile)
  setRememberedRoute(locationPathname, activeProfile)
}
```

The user is still sitting **on** the exhausted session's route (that is what the
error state is), so `routedSessionId` is truthy and the row genuinely belongs to
the profile — the guard passes and the dead id goes back into storage.

Effect 1 only re-runs when `activeProfile` / `profileReady` /
`resumeExhaustedSessionId` change. Effect 2 also depends on `sessions` and
`locationPathname`. So any ordinary re-render — a routine session-list refresh
is enough — runs **effect 2 alone**, with nothing left to undo the bad write.
Net result after the dust settles: `lastRoute` and `lastSessionId` both point at
a session that cannot be resumed, permanently.

## Reproduce

1. Open a chat, quit the app.
2. Make that session unresumable (delete it from another surface, or point the
   app at a backend where the resume rejects).
3. Relaunch → boots into the session, retries exhaust, error state appears.
4. Wait for one session-list refresh, quit, relaunch.

**Observed:** every subsequent launch reopens the same dead session. `/new`
fixes it only until the next launch.
**Expected:** a session that failed terminally is not remembered as the place to
return to.

Verified on the real app: `hermes.desktop.lastRoute.profile.default` and
`hermes.desktop.lastSessionId.profile.default` in the Electron LevelDB both held
the dead id across many restarts.

## Failing test

Added to `use-desktop-integrations.test.tsx`. The existing exhausted-cleanup
tests all render at `/`, which is why this slipped through — the real shape is
being routed **at** the dead session.

```ts
it('does NOT re-remember the exhausted session on a later re-render', () => {
  const sessions = [session({ id: 'exhausted', profile: 'default' })]

  const { rerender } = render({
    locationPathname: '/exhausted',
    profileReady: true,
    resumeExhaustedSessionId: 'exhausted',
    routedSessionId: 'exhausted',
    sessions
  })

  expect(window.localStorage.getItem('hermes.desktop.lastSessionId.profile.default')).toBeNull()

  // Plain session-list refresh: new array identity, same exhausted latch.
  // Only the persistence effect re-runs.
  rerender({
    activeProfile: 'default',
    locationPathname: '/exhausted',
    profileReady: true,
    resumeExhaustedSessionId: 'exhausted',
    routedSessionId: 'exhausted',
    sessions: [session({ id: 'exhausted', profile: 'default' })]
  })

  expect(window.localStorage.getItem('hermes.desktop.lastSessionId.profile.default')).toBeNull()
  expect(window.localStorage.getItem('hermes.desktop.lastRoute.profile.default')).toBeNull()
})
```

Without the fix: `AssertionError: expected 'exhausted' to be null`.

## Fix

Teach the persistence effect about the exhausted latch, and add the dep so it
re-evaluates when the latch changes:

```ts
if (
  routedSessionId &&
  routedSessionId !== resumeExhaustedSessionId &&
  sessionBelongsToProfile(sessions, routedSessionId, activeProfile)
) {
  setRememberedSessionId(routedSessionId, activeProfile)
  setRememberedRoute(locationPathname, activeProfile)
}
```

The suppression is scoped to the error state only — once the session resumes
normally the latch clears and it is remembered again (covered by a second test).

Locally: targeted suites 134 passed, `npm run typecheck` and ESLint clean, and
the remembered keys self-heal on the next launch of the patched build.

Happy to open a PR if useful.

## Related

- #60541 — same stranding symptom; its "the pinned id is cleared" premise is
  what this defect breaks.
- #67859 — the sibling case (remembered route is `/`), a different branch of the
  same hook.

## Environment

- Hermes desktop 0.17.0, packaged build, macOS 26.3.1 (arm64), Electron 40.10.2

kokhlo:
## Summary

The persistence effect in `useDesktopIntegrations` re-runs on every ordinary session-list refresh (its deps include `sessions` and `locationPathname`), while the cleanup effect that drops an exhausted session id only re-runs when `activeProfile` / `profileReady` / `resumeExhaustedSessionId` change. Sitting on the exhausted session's route, the persistence guard still passes (`routedSessionId` is truthy, the row still belongs to the profile) and writes the dead id straight back into `lastSessionId` / `lastRoute` — outliving the cleanup and every restart (#98467).

This adds a write barrier: while the routed session is the resume-exhausted one, the persistence effect skips both remembered-navigation writes. `resumeExhaustedSessionId` joins the effect deps so the barrier clears as soon as the id changes.

Fixes #98467

### DIFF
--- a/apps/desktop/src/app/contrib/hooks/use-desktop-integrations.test.tsx
+++ b/apps/desktop/src/app/contrib/hooks/use-desktop-integrations.test.tsx
@@ -179,6 +179,39 @@ describe('useDesktopIntegrations', () => {
     })
   })
 
+
+    it('does not re-persist an exhausted session on a session-list refresh', () => {
+      window.localStorage.setItem('hermes.desktop.lastRoute.profile.default', '/exhausted-session')
+      window.localStorage.setItem('hermes.desktop.lastSessionId.profile.default', 'exhausted-session')
+
+      const sessions = [session({ id: 'exhausted-session', profile: 'default' })]
+
+      const result = render({ profileReady: true, sessions })
+
+      // The cleanup effect drops the remembered exhausted session...
+      result.rerender({
+        activeProfile: 'default',
+        locationPathname: '/exhausted-session',
+        profileReady: true,
+        resumeExhaustedSessionId: 'exhausted-session',
+        routedSessionId: 'exhausted-session',
+        sessions
+      })
+
+      // ...and a routine session-list refresh (same sessions, ordinary re-render)
+      // must not write the dead id back into remembered navigation.
+      result.rerender({
+        activeProfile: 'default',
+        locationPathname: '/exhausted-session',
+        profileReady: true,
+        resumeExhaustedSessionId: 'exhausted-session',
+        routedSessionId: 'exhausted-session',
+        sessions: [...sessions, session({ id: 'other-session', profile: 'default' })],
+      })
+
+      expect(window.localStorage.getItem('hermes.desktop.lastSessionId.profile.default')).toBeNull()
+      expect(window.localStorage.getItem('hermes.desktop.lastRoute.profile.default')).toBeNull()
+    })
   describe('ownership validation', () => {
     it('refuses to restore a session route owned by another profile', () => {
       window.localStorage.setItem('hermes.desktop.lastRoute.profile.default', '/ai-session')
--- a/apps/desktop/src/app/contrib/hooks/use-desktop-integrations.ts
+++ b/apps/desktop/src/app/contrib/hooks/use-desktop-integrations.ts
@@ -157,13 +157,24 @@ export function useDesktopIntegrations({
     // non-overlay route (a page like /skills, or a session route) per profile.
     // Session-shaped routes require an explicit matching owner; unresolved and
     // wrong-profile rows must not replace known-safe navigation.
-    if (routedSessionId && sessionBelongsToProfile(sessions, routedSessionId, activeProfile)) {
+    // The resume-exhausted session must not be written back into remembered
+    // navigation: the cleanup effect above drops it once, but this
+    // persistence effect re-runs on every session-list refresh while its
+    // deps are unchanged — without the barrier the dead id outlives every
+    // restart and the window boots into the resume-error screen each time.
+    const exhausted = routedSessionId !== null && routedSessionId === resumeExhaustedSessionId
+
+    if (
+      routedSessionId &&
+      !exhausted &&
+      sessionBelongsToProfile(sessions, routedSessionId, activeProfile)
+    ) {
       setRememberedSessionId(routedSessionId, activeProfile)
       setRememberedRoute(locationPathname, activeProfile)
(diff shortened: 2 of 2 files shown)


Enough1122:

**98475 — Don't persist resume-exhausted session into remembered navigation**

- `apps/desktop/src/app/contrib/hooks/use-desktop-integrations.ts:15-28` adds `exhausted = routedSessionId !== null && routedSessionId === resumeExhaustedSessionId` barrier before `setRememberedSessionId`/`setRememberedRoute`. Comment: cleanup effect drops exhausted id once, but persistence effect re-runs on every session-list refresh while deps unchanged — without barrier dead id outlives restarts and window boots into resume-error screen each time. Dependency array now includes `resumeExhaustedSessionId`.

Non-blocking:
- Single-condition guard, low risk; `resumeExhaustedSessionId` appears to be from higher hook — ensure it's stable (not new object each render) to avoid unnecessary effect churn, but including it is correct. **Non-blocking**.
- No test diff supplied — manual verify that normal `routedSessionId` still persists when not exhausted. **Non-blocking**.

Verdict: Small navigation memory leak fix, prevents boot-to-error regression.

---

kokhlo:
Both points addressed: added a regression test that drives the exact failure mode — cleanup effect drops the exhausted id, then an ordinary session-list refresh rerender (new session row, deps of the cleanup effect unchanged) must not write it back. `resumeExhaustedSessionId` is a `string | null` prop from the caller, so the dep is a primitive — no object-identity churn. Pushed as 09c19cd.

Issue #53876, opened 2026-06-28
Feature Request: Make session compression warning threshold configurable


crayontomas:
## Summary

The session compression warning ("⚠️ Session compressed N times — accuracy may degrade. Consider /new to start fresh.") currently triggers after **2 compressions**, hardcoded in source:

```python
# agent/conversation_compression.py:739
if _cc >= 2:
```

## Problem

- The threshold `2` is too aggressive for long-running sessions (e.g., multi-agent orchestration, research tasks that naturally accumulate context).
- Users who are comfortable with higher compression counts get spammed with the warning on every turn after the 2nd compression.
- No way to suppress or adjust without patching source code — and `hermes update` overwrites local changes.

## Proposed Solution

Add a configurable threshold under the existing `compression` config section:

```yaml
compression:
  enabled: true
  threshold: 0.50
  target_ratio: 0.20
  warn_after_compressions: 2   # NEW: default 2 (current behavior), set to 0 to disable
```

The change in `conversation_compression.py` line 739 would read from config:

```python
warn_threshold = getattr(agent, '_compression_warn_threshold', 2)
if _cc >= warn_threshold and warn_threshold > 0:
```

## Environment

- Hermes version: 0.17.0 (editable install)
- Python 3.11
- All platforms affected (CLI, TUI, gateway)

PR #98477, opened 2026-08-30
feat(cli): add /compthreshold for session-scoped compression threshold override

gpipi95:
## Summary

Adds `/compthreshold` — a CLI command to set / query / clear a **session-scoped token threshold override** for context compression. When set, the override wins over heuristic ratio-derived triggers, so users can force compression to fire earlier (or later) for the current session without editing config.

## Motivation

- Upstream request: #53876 (configurable compression warning thresholds)
- Today compression triggers are computed purely from context-window ratios; there is no per-session knob. Long-running sessions (heavy tool output, big attachments) often need a session-specific threshold.

## Implementation

- `agent/context_compressor.py`:
  - `apply_session_threshold_override()` — checked before every ratio-derived trigger, so the override survives window re-resolution and model switches
  - `get/set/clear` session-scoped threshold helpers persisted in the session DB (`/compthreshold 350k`, `/compthreshold clear`)
- `cli.py` + `hermes_cli/commands.py`:
  - Slash-command entry point (CLI-only for now; gateway dispatch is a follow-up)
- Tests: `tests/agent/test_compthreshold_session_override.py` — 20 cases covering set/query/clear, precedence over heuristic ratios, invalid inputs, and persistence.

## Notes

- No changes to compression behavior when no override is set.
- No new dependencies.

### DIFF
--- a/agent/context_compressor.py
+++ b/agent/context_compressor.py
@@ -2307,6 +2307,8 @@ class ContextCompressor(ContextEngine):
         self._last_compression_made_progress = False
         self._summary_failure_cooldown_until = 0.0  # transient errors must not block a fresh session
         self._cooldown_persist_failed = False
+        # Session-scoped threshold override does NOT survive /new|/reset.
+        self.session_threshold_tokens = None
         self._last_summary_error = None
         self._last_compress_aborted = False
         self._last_compress_refused_would_grow = False
@@ -2519,6 +2521,13 @@ class ContextCompressor(ContextEngine):
 
     @property
     def threshold_tokens(self) -> int:
+        # Session-scoped token override (set via /compthreshold 350k) wins
+        # over every ratio-derived trigger. Checked first so the override
+        # survives window re-resolution and model switches without having to
+        # intercept every recompute path.
+        _session_tokens = self._resolve_session_threshold_tokens()
+        if _session_tokens is not None:
+            return _session_tokens
         if self._threshold_tokens is None:
             # Resolve the window FIRST (may apply the small-context floor to
             # threshold_percent as a side effect) so the percent read below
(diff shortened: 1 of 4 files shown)


Enough1122:

### PR #98477 — feat(cli): add /compthreshold for session-scoped compression threshold override

**Summary:** Adds session-scoped token override `session_threshold_tokens` on `ContextCompressor` via `/compthreshold` (CLI) with a shared `apply_session_threshold_override()` API. Override wins over per-model/global/threshold-floor, clamped to `[64K, 95% window]`, cleared on `/new`/`/reset`, and survives model switches/window re-resolution. Invalidates `_threshold_tokens`/`_tail_token_budget` on set/reset and restores `threshold_percent` from `_base_threshold_percent` so auto-lowered percent does not stick.

- `threshold_tokens` getter delegates to `_resolve_session_threshold_tokens()` (`agent/context_compressor.py:2530`) before any ratio path — single place to apply clamp `[1, 95% window]`. When `context_length` is unknown (`0`/`None`) it returns raw tokens; clamp still applies once window resolves. 64K floor is enforced only in the slash-command parser, so direct `session_threshold_tokens = 10_000` still stringifies as 10K (`test_override_below_64k_still_resolves`) — intentional split between engine (clamp) and UX (floor).
- Setter heuristic (`agent/context_compressor.py:2550`) — `if _session is not None and value < _session: self.session_threshold_tokens = value` — lets the auxiliary `auto-lower` correction (`conversation_compression.py:1804`) shrink the override, while a model-switch recompute (ratio-derived, usually ≥ override) leaves the user's setting intact. If a future recompute legitimately computes a lower value than the override for a smaller-window model, it will still be treated as auto-lower and lower the override — likely desired, but documenting the `<` rule helps.
- `on_session_reset()` nulls the override (`agent/context_compressor.py:2310`) — correct session scope. `apply_session_threshold_override("show"|"status")` and `reset` share one parse/clamp/assign path for CLI and TUI, reducing drift.
- Bare number defaults to K (`multiplier = 1_000`), so `/compthreshold 80` → 80K, `80k` and `80K` (after `.lower()`) and `1.5m` all work; comma stripping handles `80,000`. `reset` is case-folded via `arg.lower()`. Good.

**Verdict:** Well-scoped UX for #53876, with thorough 15-test coverage of override vs per-model vs floor vs clamp vs auto-lower vs reset. No blocking issues.

PR #98480, opened 2026-08-30
fix(agent): default relay completions to the progress-hook wrapper

### LINKED ISSUE #98466: [Bug]: Auxiliary retry/fallback attempts bypass the progress-hook wrapper — 18 of 24 relay call sites pass no create=, so the compression idle watchdog kills healthy calls
### Bug Description

`_create_with_progress` (added in #71508) is applied per **call site**, not at the routing seam. `_relay_sync_completion` / `_relay_async_completion` still default to a bare callback:

```python
# agent/auxiliary_client.py, main @ 2a598aad — :3558 (sync), :3586 (async)
callback = create or (lambda request: client.chat.completions.create(**request))
```

Auditing every `_relay_*_completion(...)` call on `main` today, **18 of 24 pass no `create=`**:

| line | function | `create=` |
|---|---|---|
| 5449, 5519 | `_call_fallback_candidate_sync` | yes |
| 10056, 10103 | `_call_llm_impl` (primary) | yes |
| 10838, 10866 | `_async_call_llm_impl` (primary) | yes |
| **5589, 5635** | **`_call_fallback_candidate_async`** | no |
| **5015, 5089** | **`_retry_same_provider_sync` / `_async`** | no |
| **10135, 10169, 10211, 10246, 10284, 10317, 10372** | `_call_llm_impl` — temperature-strip, max_tokens-strip, Nous stale-model heal, auth refresh, credential-pool rotation | no |
| **10884, 10915, 10957, 10991, 11028, 11060, 11109** | `_async_call_llm_impl`, same recovery paths | no |

Those attempts go out non-streaming and tick the forward-progress hook **zero** times. Two consequences, both the opposite of what #71508 intended:

1. The configured `timeout` reverts to a hard wall-clock budget instead of an inter-chunk idle timeout.
2. `gateway/run.py` session hygiene (`compression.hygiene_timeout_seconds`, default 30s) and `CompressionCommitFence.touch_progress` see total silence, so they kill a **healthy, still-generating** call at its idle deadline.

The module comment above `_aux_stream_total_ceiling` states the intended invariant — *"When a forward-progress hook is installed … the primary chat.completions attempt is upgraded"*. The word "primary" is the bug: the hook is installed per **task**, not per attempt, so every attempt for that task needs it.

That the sync fallback-candidate path already carries `create=` while its async twin does not is itself the argument for fixing this at the seam: the rule currently has to be remembered at every new call site, and it has already been missed 18 times.

### Steps to Reproduce

Call any unwrapped site against a real provider with a compression-style progress hook installed — the same shape the compressor uses:

```python
with aux.aux_progress_hook(lambda: ticks.append(time.monotonic())):
    aux._call_fallback_candidate_sync(          # unwrapped on the async twin today
        client, model, "fallback_chain[0](openrouter)",
        task="compression", messages=msgs,
        temperature=None, max_tokens=None, tools=None,
        effective_timeout=120.0, effective_extra_body={}, reasoning_config=None,
    )
```

Observed on the wire, identical prompt, unwrapped vs. wrapped:

```
unwrapped   wire: stream=None    ticks: 0
wrapped     wire: stream=True    ticks: 13   (first 0.30s, largest gap 1.69s)
```

The primary attempt with the same prompt ticks ~1300–2100 times. So the watchdog's "idle" window is, on an unwrapped route, a hard deadline.

### Expected Behavior

Every auxiliary attempt for a task with an installed progress hook streams and reports forward progress, so the idle watchdogs measure actual inactivity.

### Actual Behavior

Only wrapped call sites do. On an unwrapped route the watchdog fires at its idle deadline while the provider is still generating.

Production evidence (from an older local checkout, v0.20.1, predating current `main`, where the **sync** fallback-candidate path was still unwrapped too): the primary compression provider began returning a persistent `429` (quota exhausted until a fixed reset date). From that moment 100% of compressions ran on the silent route and none could succeed. User-visible symptom on every oversized turn:

```
⚠️ Context compression timed out after 30.0s with no output from the summary model.
```

plus `made no progress for 120.0s` on the agent path. Widening the watchdog windows does not help — the elapsed time is real wall clock on a call that is producing tokens the watchdog cannot see. Billing confirmed the calls were killed mid-flight rather than failing: spend stayed at ~0 while every attempt "timed out". On current `main` the same failure mode remains reachable through the async fallback chain and all the recovery/retry paths listed above.

### Proposed Fix

Move the wrapper from the call sites to the seam, making the invariant structural:

```python
-    callback = create or (lambda request: client.chat.completions.create(**request))
+    callback = create or (lambda request: _create_with_progress(client, request))
```

`_create_with_progress` is a byte-for-byte passthrough when no hook is installed (`not _aux_progress_active() and not force_stream`), so callers outside compression are unchanged — that is the property that makes a seam default safe.

The async side needs a small mirror, `_acreate_with_progress()`, since only the unconditional `_acreate_with_stream()` exists today, plus an `_async_client_streams_internally()` predicate, because the three internally-streaming wire adapters are separate classes in async.

Explicit `create=` at the sites that have it is kept — they additionally pass `force_stream=_provider_requires_stream(...)`, which a seam default cannot compute.

### Affected Component

`agent/auxiliary_client.py` (auxiliary completion routing) — surfaces as context compression via `gateway/run.py` session hygiene and `CompressionCommitFence`.

### Relationship to Open Work

* **#92193** (open) fixes the `force_stream` half on `_retry_same_provider_sync/_async`. Complementary: it does not cover the progress-hook half, nor the async fallback-chain candidates. A seam default would subsume its retry-path fix for the progress half; the two can land independently.
* **#71508** introduced the hook (primary only). **#71552** added stream-only providers and the stream ceiling.
* **#78981** (closed) is a different compression-hang mechanism (600s stalled-stream ceiling).

I searched open and closed issues and PRs for the progress-hook-on-unwrapped-route case and found no existing report.

### Tests

Locally verified with a `TestRelaySeamAppliesProgressWrapper` suite: sync seam streams and ticks with no explicit `create=`; sync seam unchanged (no `stream` key, same object identity) without a hook; async mirror likewise. Full `tests/agent`, clean tree vs patched: identical failure sets, the only difference being the new tests.

One existing assertion needs relaxing: `tests/agent/test_auxiliary_explicit_cancellation.py::test_isolated_provider_worker_inherits_protection_and_progress_hook` asserts `progress == ["tick"]`. Its invariant is that the hook reaches the isolated worker thread; the exact tally is incidental, and the seam also ticks on dispatch. Suggested: `assert progress and set(progress) == {"tick"}`.

I'm happy to open a PR with the patch and tests if the approach looks right.

### Environment

* Reproduced on Hermes Agent v0.20.1 (2026.8.13), install method `git`, Python 3.11.15, Ubuntu Linux (kernel 6.17)
* Code audit above is against `main` @ `2a598aad` (fetched 2026-08-30); line numbers refer to that revision
* Provider-independent (verified against OpenRouter and an OpenAI-compatible Alibaba endpoint); the defect is in the shared routing seam

### Debug Report

Not attached — this is a code-path defect with a self-contained reproduction rather than an environment issue, and a debug bundle would carry endpoint/credential context. Happy to supply specific values on request.

kokhlo:
## Summary

`_create_with_progress` was applied per call site, so auxiliary retry and fallback attempts that call `_relay_sync_completion` / `_relay_async_completion` without `create=` went out through the bare `client.chat.completions.create` — non-streaming, ticking the forward-progress hook zero times. The configured `timeout` reverted to a hard wall-clock budget and the compression hygiene watchdog (`compression.hygiene_timeout_seconds`, default 30s) killed healthy, still-generating calls at the idle deadline (#98466: 18 of 24 relay call sites pass no `create=`).

This moves the wrapper to the seam: when `create` is not supplied, both relay functions default to `_create_with_progress(client, request)`. All 18 unwrapped sites are covered at once; the 6 sites that already thread their own `create=` keep their behavior unchanged. New call sites can no longer forget the wrapper.

Behavior when no progress hook is active is byte-for-byte identical (`_create_with_progress` falls straight through to the plain create in that case), so only tasks with an installed hook change at all — and for them this restores the intended invariant from #71508: every attempt for the task streams and ticks the hook.

Fixes #98466

### DIFF
--- a/agent/auxiliary_client.py
+++ b/agent/auxiliary_client.py
@@ -3555,7 +3555,14 @@ def _relay_sync_completion(
     api_mode: str | None = None,
     create: Callable[[dict[str, Any]], Any] | None = None,
 ) -> Any:
-    callback = create or (lambda request: client.chat.completions.create(**request))
+    # The progress hook is installed per task, so every attempt for that
+    # task must go through _create_with_progress — including auxiliary
+    # retries and provider fallbacks. Defaulting at this seam covers every
+    # call site that predates the hook or forgets to thread create= through
+    # (18 of 24 today); call sites that pass their own create= keep it.
+    if create is None:
+        create = lambda request: _create_with_progress(client, request)
+    callback = create
     route = _relay_auxiliary_metadata(provider=provider, api_mode=api_mode)
     # Protected compression calls isolate only the provider callback and stream
     # aggregation.  The owning thread remains free to unwind its lease/DB
@@ -3583,7 +3590,14 @@ async def _relay_async_completion(
     api_mode: str | None = None,
     create: Callable[[dict[str, Any]], Any] | None = None,
 ) -> Any:
-    callback = create or (lambda request: client.chat.completions.create(**request))
+    # The progress hook is installed per task, so every attempt for that
+    # task must go through _create_with_progress — including auxiliary
+    # retries and provider fallbacks. Defaulting at this seam covers every
+    # call site that predates the hook or forgets to thread create= through
+    # (18 of 24 today); call sites that pass their own create= keep it.
+    if create is None:
+        create = lambda request: _create_with_progress(client, request)
+    callback = create
     route = _relay_auxiliary_metadata(provider=provider, api_mode=api_mode)
     if route is None:
         return await callback(kwargs)


Enough1122:

### PR #98480 — fix(agent): default relay completions to the progress-hook wrapper

**Summary:** `_relay_sync_completion` and `_relay_async_completion` (`agent/auxiliary_client.py:3555`) now default `create` to `lambda request: _create_with_progress(client, request)` when no caller-supplied `create=` is passed. The forward-progress hook is installed per task, so auxiliary retries and provider fallbacks that call the relay without threading `create=` through previously bypassed the wrapper (18 of 24 call sites), losing progress ticks and reverting timeout to a hard wall-clock that the compression hygiene watchdog (`compression.hygiene_timeout_seconds`, default 30s) treated as stalled.

- Preserves explicit `create=` — call sites that already pass a streaming/progress wrapper keep it. Only the `None` default gains the wrapper.
- Lambda captures `client` by closure; `create` signature `Callable[[dict], Any]` matches `_create_with_progress(client, request)` return. Symmetric fix for sync and async seams.
- Fix is at the narrow waist: one default in each relay covers all current and future call sites that forget to thread `create=` through, without touching 18 call sites individually.

**Verdict:** Minimal, correct seam fix for progress-hook starvation. No issues.

teknium1:
Landed on `main` via #106866 with your commit cherry-picked intact (authorship preserved): `fba6110cde4` — fix(agent): default relay completions to the progress-hook wrapper.

Follow-ups stacked on top in the same PR: the async twin `_acreate_with_progress` (+ `_async_client_streams_internally`) so `_relay_async_completion` and the async primary also stream/tick, and one invariant test pinning both relay defaults. Thanks @kokhlo — the per-call-site audit table made the seam fix the obvious one.

Closing as merged-via-salvage.

Issue #83626, opened 2026-08-11
Cron command guards ignore cron_mode when HERMES_EXEC_ASK is inherited


fangliquanflq:
## Problem

Cron jobs running inside a gateway or Desktop backend inherit `HERMES_EXEC_ASK=1`. When a terminal command triggers the combined approval guard, the job can return `pending_approval` even though cron runs are unattended and should be governed by `approvals.cron_mode`.

This leaves no listener capable of answering the approval request and can stall or abort otherwise valid scheduled work.

## Root cause

`check_all_command_guards()` currently evaluates the cron policy only inside the non-interactive branch guarded by `not is_ask`. An inherited `HERMES_EXEC_ASK=1` therefore makes the cron branch unreachable and routes the command into the gateway/ask approval path.

## Expected behavior

- Cron runs resolve `approvals.cron_mode` before any interactive approval path.
- `cron_mode: deny` blocks flagged commands without creating a pending approval.
- `cron_mode: approve` permits commands without waiting for a listener.
- Existing dangerous-pattern, Tirith, and `tirith_fail_open` behavior remains intact.

## Reproduction

Bind a cron session context while `HERMES_EXEC_ASK=1` and manual approvals are active, then call `check_all_command_guards()` for a flagged command. The current result has `status: pending_approval` instead of applying the cron policy.

An older implementation attempt exists in #50409, but it predates the current complete cron security branch and no longer applies cleanly to current main.

jimmywu0816:
Confirming this is still present on current `main` (`d20a8e4475`), with an additional security consequence beyond the availability symptom described here: **`approvals.cron_mode: deny` is not just unreachable — with `approvals.mode: smart` (the default), a guardian-LLM APPROVE silently executes the flagged command inside the cron job.** The "unattended ⇒ deny" guarantee degrades to "whatever the guardian LLM decides".

### Why current `main` still hits it

`check_all_command_guards` (`tools/approval.py`) still evaluates the unattended contexts only inside the non-ask branch:

```python
if not is_cli and not is_gateway and not is_ask:
    for ctx in _unattended_contexts():
        result = _unattended_deny(command, ctx)
```

With gateway-startup `HERMES_EXEC_ASK=1` inherited by in-process cron jobs, flagged commands fall through to `_human_decision`, where smart mode consults the guardian LLM first: APPROVE ⇒ executed, with `cron_mode` never consulted anywhere on the path; DENY ⇒ `pending_approval` (the stall this issue describes).

[...]

Issue #87509, opened 2026-08-16
[Bug]: API-server sessions (/v1/runs) have no cron_mode-equivalent approval exemption and stall for approvals.timeout


robbintops:
> [!NOTE]
> **Correction (2026-08-23).** Two claims in the original report below were wrong, and are struck through in place rather than deleted so [@jackulau's reply](https://github.com/NousResearch/hermes-agent/issues/87509#issuecomment-5306458957) still reads in context:
>
> 1. **#54019 was not fixed.** It was closed as a *duplicate*. The canonical webhook thread is **#37284**, still open. So there is no webhook precedent — this gap is unsolved for every listener-less platform, not just `api_server`.
> 2. **The suggested fix should not be built as written.** `approvals.api_server_mode` would collide with **#37317**, which generalises the non-cron exemption inside `_is_gateway_approval_context()`. `api_server` should opt into that instead.
>
> The report's substance is unchanged: `/v1/runs` takes the gateway approval branch with no listener, and stalls for the full `approvals.timeout`. Filing it separately still looks right — `/v1/runs` is unattended *by definition*, which is the argument for making the exemption general rather than per-platform.

**Version:** Hermes Agent v0.20.1 (2026.8.13), Linux (Hostinger `hvps-hermes-agent` image, built 2026-08-14)

## Problem

`_is_gateway_approval_context()` in `tools/approval.py` exempts cron from the blocking gateway-approval branch, with this rationale in its own docstring:

> "Cron jobs are NEVER gateway-approval contexts even when they originate from a gateway platform (cron binds HERMES_SESSION_PLATFORM via contextvars for delivery routing). Cron approvals are governed by `approvals.cron_mode` config, not interactive resolve — letting cron fall through to the gateway branch would submit a pending approval with no listener and block the job indefinitely."

**API-server sessions (`POST /v1/runs`) have the identical "no listener present" property and get no such exemption.** In v0.20.1:

```python
def _is_gateway_approval_context() -> bool:
    if _is_cron_approval_context():
        return False
    if env_var_enabled("HERMES_GATEWAY_SESSION"):
        return True
    return bool(_get_session_platform())      # api_server lands here
```

[...]

Issue #87183, opened 2026-08-15
CLI approval panel never renders: relay_runtime imports gateway.run whose module top-level setenv (HERMES_EXEC_ASK=1) hijacks approvals into the gateway path with no notify_cb


a1398394385:
## Summary

In CLI sessions, dangerous-command approvals return `{"status": "pending_approval", "approval_pending": true}` and the prompt_toolkit approval panel **never renders**. The command hangs forever with no listener. Same env-leak family as #83626 / #63183, but with a concrete writer identified.

## Root cause

`agent/relay_runtime.py::_segments_config()` contains:

```python
from gateway.run import _load_gateway_config  # late import
```

`gateway/run.py` executes these at **module top level** (not guarded by `__main__` or any gateway-only check):

```python
os.environ["_HERMES_GATEWAY"] = "1"   # ~line 1868
os.environ["HERMES_QUIET"] = "1"      # ~line 2400
os.environ["HERMES_EXEC_ASK"] = "1"   # ~line 2403
```

Any process that imports `gateway.run` — including a plain CLI/TUI session — gets these gateway-process env vars written into **its own `os.environ`** at runtime. Then `tools/approval.py:3344`:

```python
if is_gateway or env_var_enabled("HERMES_EXEC_ASK"):
```

routes approval to the gateway async path. The CLI registers no `notify_cb` for that path, so the request falls into `submit_pending()` and is silently queued — no panel, no listener, command hangs forever.

`_segments_config()` is invoked from the relay turn lifecycle (`agent/relay_runtime.py:922` in `begin_turn`, `:1082` in the compaction-flag path), which CLI sessions hit on every turn.

## Evidence

- Process starts with a clean environment, but at runtime `getenv("HERMES_EXEC_ASK")` returns `"1"` (verified by attaching lldb and reading the live environ; `psutil.environ()` is not a reliable oracle on macOS because it only reflects the startup environment).
- The poisoned `HERMES_EXEC_ASK=1` is then inherited by terminal subprocesses and persisted into the `hermes-snap-*.sh` env snapshots, self-replicating across restarts — so restarting the terminal does **not** help.
- Reproduced on `main` (as of 2026-08-16) in plain CLI mode, macOS.

## Related issues

- #83626 — Cron command guards ignore cron_mode when HERMES_EXEC_ASK is inherited (attributed to "inheritance"; the actual writer can be this import side effect)
[...]

PR #83631, opened 2026-08-11
fix(approval): prioritize cron policy over exec ask

fangliquanflq:
## Summary

- resolve unattended cron approval policy before inherited gateway/Desktop `HERMES_EXEC_ASK` routing
- preserve the complete dangerous-pattern, Tirith, and `tirith_fail_open` deny path
- cover both `cron_mode: deny` and `cron_mode: approve` with the real ContextVar plus inherited-env combination

## Problem

Gateway-hosted cron jobs inherit `HERMES_EXEC_ASK=1`. The combined command guard previously checked cron policy only under `not is_ask`, so flagged commands entered the interactive approval path and returned `pending_approval` even though no listener is present during a cron run.

This supersedes the stale implementation in #50409 while preserving its original author through commit co-authorship. It also incorporates the maintainer review requirement to retain the current full cron security branch rather than restoring the older abbreviated check.

## Testing

- `scripts/run_tests.sh tests/tools/test_cron_approval_mode.py -v` — 32 passed
- `scripts/run_tests.sh tests/cron/test_scheduler_cron_session_isolation.py -q` — 1 passed
- regression test confirmed red before the fix (returned `pending_approval`) and green after the fix
- broader `tests/tools/test_approval.py`: 94 passed, 2 unrelated Windows failures reproduced unchanged on pristine upstream `main`

Closes #83626

PR #98483, opened 2026-08-30
fix(approvals): let cron_mode govern gateway-hosted cron fires

arek-eeva:
### The problem

Cron jobs run inside the long-lived gateway process. That process exports `HERMES_INTERACTIVE=1` and `HERMES_EXEC_ASK=1` for its attended turns, and those markers are still set when an unattended cron fires.

`check_all_command_guards` reads them, so a cron fire looks interactive. It skips the deterministic `approvals.cron_mode` branch and falls through to smart or human approval instead, with nobody present to answer. Under `approvals.cron_mode: deny` the command is neither denied nor deterministic.

### The fix

Clear `is_cli`, `is_gateway` and `is_ask` once cron context is bound, so `approvals.cron_mode` governs. This mirrors the block immediately above it, which already clears `is_ask` so `single_query_mode` takes effect.

### Test

`test_cron_deny_ignores_gateway_interactive_and_ask_env_leaks` covers the case that surfaced this: a command containing the quoted text `hermes update`, fired by cron under `cron_mode: deny`, with the gateway's interactive and ask markers set.

Verified against `main` at 2a598aad1:

- test on unpatched `main`: fails, reaching the human-approval path
- with the fix: passes
- `tests/tools/test_cron_approval_mode.py`: 31 passed
- all `tests/tools/*approval*.py`: same 7 pre-existing failures before and after, 348 → 349 passed

The 7 unrelated failures are in `test_approval.py` and `test_approval_mode_parity.py` and are present on unpatched `main` in the same environment.

### DIFF
--- a/tests/tools/test_cron_approval_mode.py
+++ b/tests/tools/test_cron_approval_mode.py
@@ -475,3 +475,34 @@ class TestCronWithGatewayOrigin:
         finally:
             clear_session_vars(tokens)
 
+    def test_cron_deny_ignores_gateway_interactive_and_ask_env_leaks(self, monkeypatch):
+        """A gateway-hosted cron fire must use cron_mode even when the gateway
+        process exports its interactive/ask markers.
+
+        Production regression: a citation command containing quoted text
+        ``hermes update`` was smart-approved because HERMES_INTERACTIVE=1 and
+        HERMES_EXEC_ASK=1 skipped the cron-deny branch.
+        """
+        monkeypatch.delenv("HERMES_CRON_SESSION", raising=False)
+        monkeypatch.setenv("HERMES_INTERACTIVE", "1")
+        monkeypatch.setenv("HERMES_EXEC_ASK", "1")
+        monkeypatch.setenv("HERMES_GATEWAY_SESSION", "1")
+        monkeypatch.delenv("HERMES_YOLO_MODE", raising=False)
+        monkeypatch.setattr(approval_module, "_YOLO_MODE_FROZEN", False)
+        monkeypatch.setattr(approval_module, "_get_approval_mode", lambda: "smart")
+        monkeypatch.setattr(approval_module, "_get_cron_approval_mode", lambda: "deny")
+
+        tokens = set_session_vars(
+            platform="telegram", chat_id="123", cron_session="1"
+        )
+        try:
+            from unittest.mock import patch as mock_patch
+            with mock_patch("tools.approval._smart_approve") as smart:
+                result = check_all_command_guards(
+                    "printf %s 'hermes update'", "local"
+                )
+            assert result["approved"] is False
+            assert "cron_mode" in result["message"]
+            smart.assert_not_called()
+        finally:
+            clear_session_vars(tokens)
--- a/tools/approval.py
+++ b/tools/approval.py
@@ -4714,6 +4714,18 @@ def check_all_command_guards(command: str, env_type: str,
         # either here) — ignore it so single_query_mode actually takes effect.
         is_ask = False
 
+    # Cron jobs also run inside the long-lived gateway process, which may export
+    # HERMES_INTERACTIVE=1 and HERMES_EXEC_ASK=1 for attended turns. Those are
+    # process-wide host markers, not permissions for this unattended fire. If
+    # they leak through here, the cron skips its deterministic cron_mode branch
+    # and can reach smart or human approval with nobody present (#production
+    # regression: quoted ``hermes update`` evidence was smart-approved). Cron
+    # context owns the decision exactly as single-query context does above.
+    if _is_cron_approval_context():
+        is_cli = False
+        is_gateway = False
+        is_ask = False
+
     # Preserve the existing non-interactive behavior: outside CLI/gateway/ask
     # flows, we do not block on approvals and we skip external guard work.
     if not is_cli and not is_gateway and not is_ask:


andrexibiza (commented):
Reviewed exact head `1712e888d504a3fc74e945e3d144b74955d4f65d` against base/current `main@2a598aad1c398e95b3325a0f100f5c28efa63d12` (the PR is exactly one commit on current main). I traced the current `check_all_command_guards()` control flow, the cron ContextVar/env fallback, the full cron deny/Tirith/fail-closed branch, the existing gateway-context cron exclusion, the new regression, exact-head workflow state, and the older approval-routing lineage.

The **mechanical direction is correct**: a bound cron execution is the qualified task-local fact; process-wide `HERMES_INTERACTIVE` / `HERMES_EXEC_ASK` / `HERMES_GATEWAY_SESSION` markers are properties of the long-lived host, not authority for this unattended fire. Clearing the three derived booleans before entering the existing `cron_mode` branch is a clean way to make the qualified execution context dominate ambient host state. I do not see a correctness defect in that 12-line routing change itself.

I do see three landing blockers around ownership, proof, and exact-object acceptance.

### P1 — this is already implemented by an open canonical lineage, with the same production hunk

The live issue for this exact defect is **#83626** (`Cron command guards ignore cron_mode when HERMES_EXEC_ASK is inherited`). It explicitly requires `cron_mode` to beat inherited interactive routing, including **both deny and approve**, while preserving dangerous-pattern, Tirith, and `tirith_fail_open` behavior.

That issue already has open **#83631** by `fangliquanflq`. Its production change is materially the same hunk as this PR:

```py
if _is_cron_approval_context():
    is_cli = False
    is_gateway = False
    is_ask = False
```

and it already carries the broader proof this head is missing: inherited `HERMES_EXEC_ASK` + `HERMES_GATEWAY_SESSION` tests for **deny and approve**, the Tirith fail-closed inherited-marker case, and the documentation sentence making cron precedence explicit. It also preserves the prior implementation lineage from **#50409** by `AustinOutside` via co-authorship. #50409 is the older stale attempt on which the premise was accepted but the maintainer review correctly required preservation of the newer Tirith / `tirith_fail_open:false` branch; #83631 is the recut that incorporated that requirement.

So #98483 should not land as a second independent owner with no #83626/#50409/#83631 provenance. There are two clean outcomes:

1. close this as duplicate in favor of #83631; or
2. if the value of this PR is **current-main rematerialization** (and it is valuable that this head is directly parented by today's main), make that supersession explicit: preserve the prior contributor credit, link/close #83626, carry forward #83631's stronger tests/docs, state that #98483 supersedes the stale implementation objects, and coordinate the older PRs so only one implementation lands.

Please do not silently replace the earlier work just because this branch is fresher. The new production insight is not independent of that lineage.

### P1 — the exact regression proves only one half of the policy

This head proves `cron_mode=deny` beats leaked gateway/interactive markers, and the `smart.assert_not_called()` assertion is useful. But the contract is bidirectional: when the operator explicitly configured `cron_mode=approve`, the exact same inherited-marker state must **allow immediately without manufacturing a pending human/smart approval**. That is not covered here.

The security side of the same branch also matters. The reason #50409 was not acceptable as-is was that an early cron return could accidentally bypass the newer Tirith and `tirith_fail_open:false` protections. This implementation routes into the current full branch rather than recreating the old bad return, which is good; however the exact head should lock that property with the inherited-marker state too. #83631 already has the relevant witness.

For a superseding recut I would require, at minimum, the exact production shape for:

- `cron_mode=deny` + task-local cron ContextVar + inherited `HERMES_EXEC_ASK/HERMES_GATEWAY_SESSION` → blocked, never pending;
- `cron_mode=approve` under the same state → approved, never pending;
- Tirith content finding and/or Tirith import failure with `tirith_fail_open:false` under the same inherited state → still fail closed in cron-deny mode.

That turns the test from “this symptom stopped” into “the authority boundary is closed on both sides.”

### Landing gate — exact-head hosted acceptance has not run

For `1712e888…`, CI `33304861212`, Docker `33304860779`, and Nix `33304860778` all concluded `action_required`; CI has **0 jobs**, and the commit currently has **0 check runs**. The repeated label-rerun workflows are also `action_required`. This is not a red product signal, but it is not green evidence either. This one-commit PR has no hosted exact-head acceptance yet, so it cannot be called merge-ready until those workflows actually execute and the required checks/attribution gate are green.

### Interlocks / other side of the shape

This belongs to the broader **ambient host marker → wrong approval owner** defect class. #87183/#88281 fixes a sibling form by preventing `gateway.run` import side effects from poisoning non-gateway hosts; here the cron really *is* hosted by the gateway, so scrubbing the process globally would be the wrong repair — the task-local cron fact must win at the mutation gate. That supports this PR's local mechanism.

The other side is **listener capability**, not just cron. #87509 (`api_server`) and #37317 (`webhook`) show the same “gateway-shaped session, nobody can answer” failure on non-cron surfaces. Do **not** broaden this PR into another platform-specific unattended policy; cron already has a canonical `cron_mode`, while the general listener-less-platform decision belongs with that existing cluster. Keep this repair narrow and make the ownership graph explicit.

One structural note: this change lands around line 4714 of `tools/approval.py`, already a >4K-line authority godfile with many open collision surfaces. If #98483 becomes the current-main convergence object, it should be the *last* duplicate policy increment here, not another parallel owner; rebase onto any accepted approval extraction that lands first rather than recreating the seam again.

The core diagnosis and the task-local-over-ambient-state framing are good work. The strongest path is to use this fresh current-main object to **converge** the older accepted lineage, not fork it. Once provenance, bidirectional/security proof, and exact-head CI are closed, this is a crisp fix.

TurgutKural (commented):
Verified the mechanism behind this fix with a local reproduction against current `main`:

- A gateway-hosted cron fire with the cron marker bound, but `HERMES_EXEC_ASK=1` (or `HERMES_INTERACTIVE=1`) present in process env, skips the deterministic `cron_mode` branch in `check_all_command_guards` and falls through to the human gate, where it strands as `pending_approval` with nobody present to answer. This matches the production symptom exactly (a scheduled `hermes update` that never runs).
- With this patch's logic applied, the same fire stays on the `cron_mode` branch: deterministically denied under `deny` without consulting the smart gate, and approved under `approve` the same way.

Two merge-blockers to note:

1. The diff has drifted against current `main` — the `tools/approval.py` hunk no longer applies (the function moved). Needs a rebase.
2. The new test references `tools.approval._smart_approve`, which no longer exists (renamed to `_smart_gate`), and patches `_get_approval_mode` / `_get_cron_approval_mode` / `_YOLO_MODE_FROZEN` on `tools.approval`, where they now live on `tools.approval_context` (mode getters) and `tools.approval` (YOLO flag) respectively. After updating those targets, the whole test file passes locally.

One suggestion: the added test only covers `deny` mode. The same leak also strands `approve`-mode fires — arguably the higher-impact case, since a scheduled update silently never runs instead of failing loudly. A mirror test asserting `approved is True` with the smart gate unconsulted, under `approve` + leaked ask marker, would close that gap.

The diagnosis in the body (process-wide host markers vs per-fire unattended policy) is accurate. Thanks for the fix.

Issue #98479, opened 2026-08-30
P0: preserve tool whitelist authorization across worker dispatch


JoaoMarcos44:
## P0 scope

This issue tracks the P0 items from the unified forensic report only. P1 follow-ups are intentionally out of scope for this change.

## Problem

Two independent P0 safety paths needed deterministic coverage:

1. `execFileNoThrow` timeout handling must settle even when a native child or daemon grandchild keeps the subprocess stdio descriptors open. Without that contract, a clipboard/process helper can remain pending after the timeout and stall its caller.
2. Plugin tool-whitelist authorization must remain active when a tool call is dispatched through a worker thread. The whitelist was stored in `threading.local()`, while worker execution uses context propagation; an unwrapped worker thread therefore observed no whitelist and could silently bypass the intended deny check.

## Root cause

- The timeout production fix is already present on `main` through the merged upstream work in #93355. This issue preserves that behavior with a cross-platform regression test rather than reworking the settled implementation.
- The plugin whitelist state was thread-local instead of context-local. `agent/tool_executor.py` propagates the parent execution context into worker calls, so the authorization state must be held in `contextvars.ContextVar`.

## Proposed P0 implementation

- Replace the plugin whitelist's thread-local storage with `ContextVar` state while preserving default/custom denial messages and clear/reset behavior.
- Add a worker-context regression test proving a denied tool remains denied inside a context-wrapped worker.
- Add a cross-platform native-child timeout regression test proving `execFileNoThrow` returns the timeout code even while child stdio remains open.
- Document scope, evidence, security invariants, validation, and the explicit P1 boundary.

## Acceptance criteria

- A tool outside the active whitelist is denied in a context-wrapped worker thread.
- Custom denial formatting remains supported and clearing the whitelist restores the default behavior.
- A timed-out native child that keeps stdio open does not leave `execFileNoThrow` pending.
- Focused Python and TypeScript tests pass through the repository's supported test commands.
- No P1 implementation is included.

## References

- Upstream timeout fix: #93355.
- The implementation PR for this issue will be linked here after it is opened.

andrexibiza:
Topology reconciliation: #98485 is the active implementation carrier for this issue; #93355 remains the already-merged owner of the underlying `execFileNoThrow` production timeout fix.

[...]

PR #93148, opened 2026-08-23
fix(tui): settle execFileNoThrow on timeout even when a daemon holds stdio

liuhao1024:
## What does this PR do?

Guarantees `execFileNoThrow()`'s returned promise settles on timeout in the default (non-`resolveOnExit`) path, fixing a forever-hang on the TUI clipboard path.

Today the timeout handler only calls `settle(124)` when `resolveOnExit: true`. The default path waits for the child's `'close'` event, which requires **every** inherited stdio handle to close — so when a spawned tool forks a daemonized grandchild that keeps the pipes open (the exact scenario `resolveOnExit` was added for, e.g. `wl-copy`-style double-forks), `'close'` never fires, and the timeout's SIGTERM only reaches the direct child, leaving the grandchild alive and the **promise pending forever**. The production clipboard path (`setClipboard()` → `tmuxLoadBuffer()` → the `osc.ts` spawn that does not pass `resolveOnExit`) hits exactly this shape (#93134). The repo already carried an `it.skip`'d regression test documenting the hang.

The fix: call `settle(124)` unconditionally in the timeout handler. The existing `settled`-guard makes it a no-op when the child's own `'exit'`/`'close'` won the race — normal timeout behavior is byte-for-byte unchanged — and in the daemon case it becomes the only exit, returning the same `124` the `'close'` path would have produced.

## Related Issue

Fixes #93134

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `ui-tui/packages/hermes-ink/src/utils/execFileNoThrow.ts` — the timeout handler now settles with `124` unconditionally instead of only under `resolveOnExit`; comment explains the daemon-grandchild mechanics.
- `ui-tui/packages/hermes-ink/src/utils/execFileNoThrow.test.ts` — un-skipped the documented-hang regression test and hardened it: the daemon script's sleeper now lives 30s so it genuinely outlives the timeout and vitest's own 5s test timeout (before the fix the test fails by timing out, exactly as the old skip-comment predicted), plus an elapsed bound (<2s) pinning the settle to the timeout, not the sleeper's death.

## How to Test

`cd ui-tui && npx vitest run packages/hermes-ink/src/utils/execFileNoThrow.test.ts` — should pass (5 tests).

[...]

Issue #93134, opened 2026-08-23
TUI clipboard path can leave execFileNoThrow pending after timeout


JoaoMarcos44:
## Problem

`execFileNoThrow()` can return a Promise that never settles when the immediate child exits but a daemonized grandchild keeps inherited stdio pipes open.

## Evidence

- Current code: `ui-tui/packages/hermes-ink/src/utils/execFileNoThrow.ts:67-79`.
- The timeout calls `settle(124)` only when `options.resolveOnExit` is true.
- The default path waits for `close`, which requires all stdio handles to close.
- The production call path `ink.tsx:1633 -> setClipboard() -> tmuxLoadBuffer() -> osc.ts:201` does not pass `resolveOnExit`.
- A runtime control-flow reproduction reported `WATCHDOG_NEVER_SETTLED` after the external 5-second watchdog in the default mode, while the exit mode settled.
- The repository already contains a skipped regression test documenting this hang in `execFileNoThrow.test.ts`.

The source was rechecked on current `main` at `0159b51f2b1cdaa8fcf65d181fc0527692724fae`; the reported control flow is still present.

## Expected behavior

The declared timeout must settle the Promise with code `124` regardless of whether the caller waits for `exit` or `close`.

## Acceptance criteria

- Settle unconditionally from the timeout handler, protected by the existing settled guard.
- Pass `resolveOnExit: true` from `tmuxLoadBuffer()`, which only consumes the exit code.
- Reactivate the skipped regression test and add coverage that the default mode cannot remain pending beyond its timeout.
- Verify the non-test `execFileNoThrow()` call sites and run the focused Vitest file.

## Scope

This is a TUI process-lifecycle bug, not a security issue. It does not change authentication, authorization, secret handling, or a security boundary.

## Related pull request

Fixes https://github.com/NousResearch/hermes-agent/pull/93209

## Validation update — current base

The implementation was re-read on current `origin/main` at `0159b51f2b1cdaa8fcf65d181fc0527692724fae` and exercised with the real exported function on Windows:

- default mode: `WITHOUT resolveOnExit -> 124 after 522ms`;
- exit mode: `WITH resolveOnExit -> 0 after 162ms`.

The focused Vitest file remains POSIX-gated on native Windows, while the TUI TypeScript typecheck passed. The runtime harness therefore supplies the host-specific evidence for the Promise lifecycle without claiming that the POSIX fixture ran on Windows.

PR #98485, opened 2026-08-30
fix: preserve P0 tool safety across worker dispatch

JoaoMarcos44:
## Summary

This PR completes the P0 tool-safety implementation for worker dispatch.

- Fixes plugin tool-whitelist state so authorization survives execution in context-propagating worker threads.
- Adds a regression test proving a disallowed tool remains denied inside a wrapped worker.
- Adds a cross-platform regression test for the existing clipboard subprocess timeout fix when a native child keeps stdio open.
- Keeps P1 work explicitly out of scope.

## Root cause

The plugin whitelist was stored in `threading.local()`. The worker execution paths intentionally propagate the current execution context, but a thread-local value is not part of that context. As a result, a worker could observe no whitelist and silently bypass the deny check.

The clipboard timeout production fix was already merged on `main` in #93355. Its regression surface needed a test that uses a native child which remains alive with inherited stdio, the condition that previously prevented the timeout promise from settling.

## Implementation

- Replaced the plugin whitelist's thread-local storage with `contextvars.ContextVar` values while preserving default/custom denial formatting and clear semantics.
- Added a worker-context authorization regression test.
- Added a native-child timeout regression test against the current `execFileNoThrow` implementation.

## Verification

- `scripts/run_tests.sh tests/hermes_cli/test_plugins.py -q`
- `npm test -- --run packages/hermes-ink/src/utils/execFileNoThrow.test.ts`
- `git diff --check`

The Python test command ran through Hermes' CI-parity test wrapper with credential isolation, UTC, and C.UTF-8. The TypeScript test exercises the real helper and native subprocess path.

## Scope and security

This PR is P0-only. It does not implement P1 items. The authorization fix is fail-closed: a worker inherits the caller's whitelist through the existing context propagation path; an absent whitelist remains the unrestricted/default mode, and a non-member tool is denied rather than executed.

Refs #98479

## Infographic :

hermes-p0-pr-98485


andrexibiza (commented):
Reviewed exact head `e8dc3a9a5bb852a6c957d1eb183491ef24792a05` against exact current `main@2a598aad1c398e95b3325a0f100f5c28efa63d12` (the PR is one commit directly on current main). I read the full diff, the existing worker-context carrier, the historical background-review safety chain, #98479, and the merged timeout repair in #93355. Exact-head CI is fully green: [CI 33304968843](https://github.com/NousResearch/hermes-agent/actions/runs/33304968843), [Docker 33304968390](https://github.com/NousResearch/hermes-agent/actions/runs/33304968390), and [Nix 33304968407](https://github.com/NousResearch/hermes-agent/actions/runs/33304968407); the CI run includes green Python/e2e, Windows/macOS, blocking ruff/footguns, contributor attribution, and `All required checks pass`.

The production direction is right: `tools/thread_context.py` deliberately propagates `contextvars.copy_context()` into worker execution, so moving the whitelist carrier out of `threading.local()` is the correct mechanism for preserving the caller's authorization state across that boundary. I do have two P1 verification blockers before I would treat the P0 class as closed, plus one landing/provenance gate.

**P1 — the new worker test proves the helper, not the vulnerable dispatch path.**

`tests/hermes_cli/test_plugins.py::test_whitelist_propagates_through_context_wrapped_worker` manually creates a thread around `propagate_context_to_thread(worker)` and calls `get_pre_tool_call_block_message()` directly. That demonstrates the `ContextVar` binding crosses *a* wrapped thread, but it never executes a real model tool through the worker-dispatch path that #98479 says was bypassing the guard. A regression in `agent/tool_executor.py` that submits one worker without `propagate_context_to_thread`, or evaluates the pre-tool gate outside the copied context, would still leave this test green.

For a P0 authorization repair, the acceptance witness needs to cross the same public/internal seam as the defect: set a restrictive whitelist, invoke an actually-dispatched disallowed tool through the worker executor (including the concurrent path if that is the path at issue), and prove the tool body is never reached and the denial result is returned. Keep the helper-level test too—it is useful and precise—but it cannot be the terminal proof for the security invariant.

**P1 — the new `execFileNoThrow` test does not reproduce the #93134/#93355 failure shape.**

The added test launches `process.execPath -e 'setTimeout(() => {}, 30_000)'`. At the 200 ms timeout, `execFileNoThrow` terminates that *same direct child*. Once that child exits, it releases its stdio handles, so the ordinary `close` path can settle. The historical bug was different: the immediate child exited while a **surviving descendant** still held the inherited stdio pipes open, so `close` never arrived. The merged #93355 regression explicitly builds that shape with `fake-daemonizer-long.sh` (`sleep 30 &; ...; exit 0`) and documents why the pre-fix promise hangs.

That means this new cross-platform test can be green without exercising the condition it claims to lock down. Please either remove it as redundant or make it a real portable descendant witness: the immediate Node child should spawn a longer-lived child inheriting the captured stdio and then exit, and the assertion should prove the fixed helper settles `124` before the descendant closes those handles. The useful teeth check is that the witness fails/hangs against the pre-#93355 conditional-settle implementation and passes against current main.

**Landing/provenance + ownership gate.**

This whitelist is not new isolated machinery. The origin chain matters because this PR repairs the mechanical safety boundary created to close the background-review side-effect class: #15204 (reported by `xlionjuan`) → #16001 / merged salvage #16569 (`luyao618`, restricted review toolsets) → #17276 (`WorldWriter`, runtime-whitelist/cache-parity design) + #25427 (`simpolism`, cache-parity refinement) → merged convergence #25434, which introduced `set_thread_tool_whitelist` / `clear_thread_tool_whitelist` and explicitly claimed preservation of the #15204 mechanical deny boundary. #98485 should preserve that lineage in its PR/issue record rather than presenting the ContextVar conversion as provenance-free new P0 ownership. The timeout half already points at #93355, whose own body preserves the #93148/#93232 salvage lineage.

There is also a current merge-order collision: open #98385 (`54c9c3875a994a43defac84fd1c4ddbeab74ac37`) modifies both `hermes_cli/plugins.py` and `tests/hermes_cli/test_plugins.py` for a different concurrent plugin-hook defect. It is **adjacent, not duplicate** work, but whichever lands second needs a current-main rebase and fresh exact-head CI. No silent overwrite of either concurrency contract.

Finally, this change adds new security-policy state at roughly line 6563 of `hermes_cli/plugins.py`, already far beyond the repository's 2K boundary. This is exactly the kind of authority state that benefits from a bounded owner: put the whitelist carrier/state operations behind a small dedicated seam and leave `plugins.py` as wiring rather than growing another security primitive inside the godfile.

One smaller artifact-truth item: `RELATORIO_P0_IMPLEMENTACAO.md` is not ready to be a permanent repository record. It still contains `PREENCHER_COM_URL_*` / `PREENCHER_COM_SHA_DO_COMMIT` placeholders, unchecked publication/CI steps, and claims `assets/infografico-p0-tool-safety.png` is in the changed-file set when the live PR has only four files and no PNG. Please remove the work report from the landing diff or make every receipt in it match the exact object.

The core ContextVar correction is good work, and the exact head is clean and fully green. Tightening the two witnesses and preserving the ownership/provenance boundary would make this a real closure of the P0 shape rather than a plausible fix with proof one layer short.

andrexibiza (commented):
Re-reviewing moved head `7e40f00d64c39f0730e257aae4de1ffa4c03f242` against current `main@4f22543509d1b91dc45bcb369447126c5eb14fb7`.

I re-read the full four-file diff, the current worker-context carrier, #98479, and the surviving commit train. The two substantive verification blockers from review 5060490532 are still present unchanged:

1. `tests/hermes_cli/test_plugins.py::test_whitelist_propagates_through_context_wrapped_worker` still proves only `propagate_context_to_thread()` + a direct call to `get_pre_tool_call_block_message()`. It never crosses the actual `agent/tool_executor.py` worker-dispatch seam and never proves a disallowed tool body is not reached. A regression that drops context propagation at the real executor boundary would still leave this test green. For a P0 authorization fix, the terminal witness needs to dispatch an actually denied tool through the real worker executor and assert both the denial result and zero tool-body execution.

2. The new `execFileNoThrow` test still starts only one direct Node child (`setTimeout(..., 30_000)`) and times out that same child. It still does not reproduce the #93134/#93355 failure shape where the immediate child exits while a surviving descendant keeps inherited stdio open. This witness can remain green without exercising the historical hang. Either remove it as redundant with #93355 or make the portable test spawn a descendant that inherits stdio, let the immediate child exit, and prove the helper settles before the descendant releases the pipe.

The moved-head release train also fails the exact-object gate. There are four surviving commits:

- `e8dc3a9a5bb852a6c957d1eb183491ef24792a05`: CI 33304968843, Docker 33304968390, Nix 33304968407 — success.
- `ea439f8a01222008d4cac81d5860bafbf763eac2`: no pull-request-triggered workflow runs found.
- `e84f72eda00266158070afb2ba017260c85b9bba`: CI 33308362299 — failure; Docker 33308361951 and Nix 33308361913 — success.
- `7e40f00d64c39f0730e257aae4de1ffa4c03f242`: CI 33313381249, Docker 33313380886, Nix 33313380903 — success.

Head-green is therefore not train-green.

Current main is 31 commits ahead of merge base `26350357d76e4508c8df9304a3374bdc5a6f6220`. The current-main-only FILE-LIST is path-disjoint from all four submitted paths, so there is no new main-side file collision; this is a stale-object/reproof gate. The separate open #98385 collision in `hermes_cli/plugins.py` and `tests/hermes_cli/test_plugins.py` still exists and remains adjacent rather than duplicate.

The ownership/artifact gates also remain: the authority carrier is still added inside `hermes_cli/plugins.py` around line 6563 instead of a bounded sub-2k owner, and `RELATORIO_P0_IMPLEMENTACAO.md` still claims `assets/infografico-p0-tool-safety.png` is versioned even though the live PR file list has no PNG and its own checklist still records remote CI / exhaustion review as unfinished.

Required completion remains: fix the two behavioral witnesses at the real failure seams, move the whitelist carrier into a bounded owner, make the repository report match the exact landing object (or remove it), recompose on current main without losing #98385's adjacent contract, and reacquire CI/Docker/Nix for every surviving commit. I am not releasing the prior blockers on this head.

teknium1:
Salvaged into #119747 with your fix commit cherry-picked intact (credit preserved): the `ContextVar` carrier is unchanged; the out-of-scope `execFileNoThrow` test and the report file were dropped, and the real-seam witness the review asked for was added (`tests/agent/test_tool_whitelist_worker_hop.py` dispatches a denied tool through `_run_sequential_tool_execution_middleware` and asserts the body never runs). A live probe confirmed the bypass on main (`allowed=None` at dispatch on the worker thread). Leaving this open for Teknium's call on which to land.

PR #91578, opened 2026-08-21
fix(skills): declarative credential_destinations in Skills Guard

liuhao1024:
## What does this PR do?

Gives Skills Guard a declarative credential-to-host policy so an expected secret-to-service relationship stops reading as exfiltration. Today the static scanner flags every read of a secret-shaped env var as CRITICAL exfiltration with no way to express "this credential belongs to this service" — which left the first-party `blocked-page-recovery` skill **uninstallable from the Hub** (DANGEROUS verdict for reading `JINA_API_KEY` and sending it to `r.jina.ai` only, #91569; `--force` correctly does not override a DANGEROUS community verdict).

A skill may now declare, in SKILL.md frontmatter:

```yaml
credential_destinations:
  JINA_API_KEY: [r.jina.ai]
```

A secret-related exfiltration finding downgrades to a low `[declared]` finding — visible in the report for auditability — ONLY when **both** hold:

1. the matched line names a secret present in the declaration, AND
2. every network host in the finding's scope (the enclosing `def`/`class` block for Python files, else a ±3-line window) is in that secret's declared destination list.

Sending the same credential to an undeclared host keeps the original severity (still DANGEROUS); an undeclared secret is not exempted; a malformed or missing declaration exempts nothing. `SCANNER_VERSION` bumps to `skills-guard-v2` so cached verdicts refresh. This implements the issue's preferred remediation (option 1) without weakening detection for undeclared hosts (option 4's concern), and no `--force` override or global allowlist is involved.

The `blocked-page-recovery` skill ships the declaration for its single credential/host pair; its archive-only routes carry no credentials and are unaffected. Runtime without `JINA_API_KEY` remains supported as before.

## Related Issue

Fixes #91569

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `tools/skills_guard.py`: Add `_load_credential_destinations` (frontmatter parse, fail-closed on any error/shape mismatch) and `_exempt_declared_credential_use` (scope-level host validation: def/class block for `.py`, ±3-line window otherwise; downgrade to low `[declared]` only when the scope's host set ⊆ the secret's declared list). Wire both into `scan_skill` before the verdict; bump `SCANNER_VERSION` to `skills-guard-v2`.
- `skills/research/blocked-page-recovery/SKILL.md`: Declare `credential_destinations: {JINA_API_KEY: [r.jina.ai]}` with a comment stating the invariant.
[...]

Issue #91569, opened 2026-08-21
blocked-page-recovery cannot be installed from Hub: Skills Guard classifies declared JINA_API_KEY use as DANGEROUS


PromptSudo:
**Hermes version:** v0.20.5 — revision `fcbd1076a93841fa88855acce810e342a5b78101`

**Exact installation command:**

```bash
hermes -p ali skills install NousResearch/hermes-agent/skills/research/blocked-page-recovery --category research --yes
```

**Source:** Official NousResearch repository ([NousResearch/hermes-agent](https://github.com/NousResearch/hermes-agent)), installed through Hermes Hub with `trust_level: community` and a fresh Skills Guard (`skills-guard-v1`) scan.

**Result:** Installation was **BLOCKED**. The security scan returned verdict **DANGEROUS** with three findings. No runtime files landed in the profile, no Hub-lock entry was created, and the quarantine staging area was cleaned.

Findings reported by Skills Guard:

- CRITICAL exfiltration — `SKILL.md:97` — a curl example sends `Authorization: Bearer $JINA_API_KEY` to `https://r.jina.ai/...`
- CRITICAL exfiltration — `scripts/recover_page.py:167` — `key = os.environ.get("JINA_API_KEY")`
- HIGH exfiltration — `scripts/recover_page.py:167` — the same environment-variable read

## Expected credential use

The skill's Jina Reader route intentionally reads `JINA_API_KEY` from the environment:

```python
def try_jina(url: str, timeout: int) -> dict | None:
    key = os.environ.get("JINA_API_KEY")
    if not key:
        return None
    status, body = _fetch(
        "https://r.jina.ai/" + url, timeout, headers={"Authorization": f"Bearer {key}"}
    )
```

The only destination receiving `JINA_API_KEY` is `https://r.jina.ai/`. Other archive services are contacted without that credential.

Runtime operation without `JINA_API_KEY` is supported because the Jina route returns `None` and is skipped when the variable is unset. However, Hub installation remains blocked even when the key is absent because Skills Guard performs static analysis of the packaged source.

This is an expected secret-to-service relationship, but the current scanner and skill metadata cannot express that a specific credential may be sent only to a specific declared host.

## No workaround applied

- `--force` correctly does not override a DANGEROUS community verdict.
- Scanning was not disabled.
- Trust metadata was not altered.
- The skill was not patched or copied from another profile.
- `JINA_API_KEY` was not configured.
- No runtime files were installed.

Current `main` remains affected.

[...]

PR #98489, opened 2026-08-30
fix(skills): ship blocked-page-recovery's Jina example in a shape our own scanners accept

### LINKED ISSUE #98474: Bundled skill research/blocked-page-recovery trips the exfil_curl context scanner — silently dropped from the prompt on every fresh v0.20.6 install
**Hermes version:** v2026.8.27 (app 0.20.6), Docker image, single profile, single operator.

v2026.8.27 ships a new stock skill, `research/blocked-page-recovery/SKILL.md`, whose Jina Reader example

```bash
curl -s -H "Authorization: Bearer $JINA_API_KEY" "https://r.jina.ai/{URL}"
```

matches the `exfil_curl` pattern in `tools/threat_patterns.py` at context scope. The prompt builder therefore drops the ENTIRE skill file from the prompt with only a `logger.warning` — the product ships a skill that its own loader refuses to serve, on every installation, with no user-visible error.

## Reproduce on a clean v2026.8.27 container

```bash
python3 - <<'PY'
import sys; sys.path.insert(0, "/opt/hermes")
from tools.threat_patterns import scan_for_threats
text = open("/opt/data/skills/research/blocked-page-recovery/SKILL.md").read()
print(scan_for_threats(text, scope="context"))   # -> ['exfil_curl']
PY
```

The scanner itself is unchanged — `tools/threat_patterns.py` is byte-identical between v2026.8.3 and v2026.8.27; what changed is the new bundled file. The other four research skills new in this window (arxiv, competitor-news-monitor, grounded-citations, llm-wiki) scan clean.

## Prior art — related but distinct surfaces

- #91569 reports the **same skill and same line** blocked by Skills Guard on a **Hub install** (loud failure, verdict shown to the user). This issue is the **bundled copy** that every fresh install already contains being dropped at **prompt-build time** — the silent variant. Fixing the skill file resolves both.
- #98322 (merged 2026-08-30) anchors the var-name suffix and drops mid-name `API` from the alternation — but it does **not** resolve this instance: `$JINA_API_KEY` *ends in* `KEY`, so `exfil_curl` still matches. The bundled skill remains self-blocking even after that fix.

## Suggested fixes (both probably worth doing)

1. **Reword the shipped skill.** Moving the header to its own line clears the pattern — under both the old and the post-#98322 regex — and keeps the skill fully functional:

   ```bash
   JR_AUTH="Authorization: Bearer $JINA_API_KEY"
   curl -s -H "$JR_AUTH" "https://r.jina.ai/{URL}"
   ```

   (Confirmed on v2026.8.27: after this edit `scan_for_threats(..., scope="context")` returns `[]`.)

2. **Lint bundled context files with the project's own scanner in CI.** Any `SKILL.md` (or other context-scope file) that ships in the image should fail the build if `scan_for_threats(text, scope="context")` is non-empty — otherwise every future skill addition can regress the same way, and downstream users get no error to notice. The same CI gate pointed at `tools/skills_guard.py` would have caught #91569 before release too.

Found the morning after our 0.20.6 bump by a daily watcher that scans every context file with the container's own scanner; without it the skill would simply never have loaded, indistinguishable from not being installed.

liuhao1024:
## What does this PR do?

The bundled `blocked-page-recovery` skill shipped its Jina Reader example with `$JINA_API_KEY` interpolated on the same line as the `curl` command:

```bash
curl -s -H "Authorization: Bearer $JINA_API_KEY" "https://r.jina.ai/{URL}"
```

That single line matches the `exfil_curl` pattern in the shared threat-pattern library at `context` scope, so the stock file trips Hermes' own scanner on every fresh install (#98474) — and the same line makes Skills Guard rate the bundled skill verdict `dangerous` with a critical `env_exfil_curl` finding, so a Hub reinstall of the product's own skill is refused (#91569). Moving the header into its own variable line clears both scanners without changing what the command does:

```bash
JR_AUTH="Authorization: Bearer $JINA_API_KEY"
curl -s -H "$JR_AUTH" "https://r.jina.ai/{URL}"
```

This is a content fix only — no scanner/guard logic is touched. Relation to #91578 (declarative `credential_destinations` in Skills Guard): that PR is the general mechanism so community skills can legitimately carry credentials to declared hosts; this PR fixes the bundled file itself so the stock distribution stops tripping the scanners today. They are complementary and touch different hunks (frontmatter vs. the bash example).

## Related Issue

Fixes #98474 (also clears the bundled-skill symptom reported in #91569; the declaration mechanism in #91578 remains valuable for third-party skills)

## Type of Change

## Changes Made

- `skills/research/blocked-page-recovery/SKILL.md`: split the Authorization header onto its own `JR_AUTH` variable line so the curl example no longer interpolates the secret env var on the command line (functional behavior unchanged)
- `tests/tools/test_bundled_skills_self_scan.py`: fail-closed anchors asserting the bundled skill stays clean under the shared context-scope scanner (#98474) and reinstallable via Skills Guard (#91569)

## How to Test

1. `pytest tests/tools/test_bundled_skills_self_scan.py -q` — Observed result: 2 passed.
2. Without the fix (revert just the SKILL.md hunk): same command — Observed result: 2 failed (context scan returns `['exfil_curl']`; `should_allow_install` returns False with verdict `dangerous`), confirming both anchors fail closed.
3. Direct scanner check against the shipped file:
   ```python
   from tools.threat_patterns import scan_for_threats
   text = open("skills/research/blocked-page-recovery/SKILL.md").read()
   print(scan_for_threats(text, scope="context"))  # -> [] after this PR (was ['exfil_curl'])
   ```
4. Adjacent suites unaffected: `pytest tests/tools/test_skills_guard.py tests/tools/test_threat_patterns.py -q` — Observed result: 64 passed.

### Code

- [x] I've run `pytest tests/ -q` and all tests pass (targeted suites for this change: `tests/tools/test_bundled_skills_self_scan.py`, `test_skills_guard.py`, `test_threat_patterns.py` — all green)

- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (pure content rewording, POSIX-shell example unchanged in semantics)

## Screenshots / Logs

Scanner verdict before → after (Skills Guard over the bundled skill dir):

```
before: verdict=dangerous allowed=False  (critical env_exfil_curl)
after:  verdict=safe      allowed=True   (only informational os.environ.get finding remains)
```

### DIFF
--- a/skills/research/blocked-page-recovery/SKILL.md
+++ b/skills/research/blocked-page-recovery/SKILL.md
@@ -94,7 +94,8 @@ returns markdown. Anonymous access is dead (401 → Turnstile); a key is
 required:
 
 ```bash
-curl -s -H "Authorization: Bearer $JINA_API_KEY" "https://r.jina.ai/{URL}"
+JR_AUTH="Authorization: Bearer $JINA_API_KEY"
+curl -s -H "$JR_AUTH" "https://r.jina.ai/{URL}"
 ```
 
 Handles JS SPAs that archives can't. Skip this route entirely when the env
--- a/tests/tools/test_bundled_skills_self_scan.py
+++ b/tests/tools/test_bundled_skills_self_scan.py
@@ -1,0 +1,41 @@
+"""Bundled skills must stay clean under Hermes' own scanners.
+
+Stock skills ship in every install, so a SKILL.md whose curl example matches
+``exfil_curl`` is a self-inflicted outage: the shared threat-pattern library
+flags it at context scope (#98474) and Skills Guard refuses the very same
+file on a Hub install (#91569). These anchors fail closed — re-introducing
+a credential-interpolating one-liner into the bundled file turns them red.
+"""
+
+from pathlib import Path
+
+from tools.skills_guard import scan_skill, should_allow_install
+from tools.threat_patterns import scan_for_threats
+
+PROJECT_ROOT = Path(__file__).resolve().parents[2]
+BUNDLED_SKILL = PROJECT_ROOT / "skills" / "research" / "blocked-page-recovery"
+
+
+def test_bundled_blocked_page_recovery_passes_context_scan():
+    """The bundled SKILL.md must not trip the shared context-scope scanner.
+
+    The Jina Reader example used to interpolate ``$JINA_API_KEY`` on the same
+    line as ``curl``, matching ``exfil_curl`` (tools/threat_patterns.py) on
+    every fresh install (#98474). Moving the header into its own variable
+    line clears the pattern while keeping the command functional.
+    """
+    text = (BUNDLED_SKILL / "SKILL.md").read_text(encoding="utf-8")
+    assert scan_for_threats(text, scope="context") == []
+
+
+def test_bundled_blocked_page_recovery_installable_from_hub():
+    """Skills Guard must allow reinstalling the bundled skill from the Hub.
+
+    The same ``exfil_curl`` line made the stock skill verdict DANGEROUS
+    under ``scan_skill`` (#91569) — the product rejected its own bundled
+    content. Only informational findings (e.g. ``os.environ.get`` key reads
+    in the script) may remain.
+    """
+    result = scan_skill(BUNDLED_SKILL, source="community")
+    allowed, reason = should_allow_install(result)
+    assert allowed, f"bundled skill must reinstall cleanly: {reason}"


Enough1122:

### PR #98489 — fix(skills): ship blocked-page-recovery's Jina example in a shape our own scanners accept

**Summary:** The bundled `blocked-page-recovery` SKILL.md's Jina Reader example interpolates `$JINA_API_KEY` on the same line as `curl -H "Authorization: Bearer $JINA_API_KEY"`, which matches the `exfil_curl` pattern in `tools/threat_patterns.py` at `context` scope. Splits into `JR_AUTH="Authorization: Bearer $JINA_API_KEY"` + `curl -s -H "$JR_AUTH"` so the file no longer self-flags on fresh install (#98474) nor fails Skills Guard (`scan_skill` DANGEROUS) on Hub reinstall (#91569). Adds `tests/tools/test_bundled_skills_self_scan.py` with two anchors: `scan_for_threats(text, scope="context")==[]` and `scan_skill(..., source="community")` allowed.

- Semantic is preserved — header value unchanged, just indirected through shell variable. Functional for copy-paste use.
- Assignment line `JR_AUTH="Authorization: Bearer $JINA_API_KEY"` itself contains the token variable but is not a `curl` invocation, so it correctly avoids the `exfil_curl` rule which requires `curl.*$VAR` co-occurrence.
- Tests pin the bundled file path (`skills/research/blocked-page-recovery/SKILL.md`) and both scanner surfaces — reintroducing a credential-interpolating one-liner fails closed.

**Verdict:** Correct self-inflicted outage fix. No issues.

liuhao1024:
Thanks for the review! Confirming the copy-paste semantics are unchanged — the shell variable is pure indirection, the header value reaching Jina is identical. The two test anchors are intentionally fail-closed on both scanner surfaces (`scan_for_threats` for the fresh-install self-scan and `scan_skill` for Hub reinstalls), so reintroducing a credential-interpolating one-liner breaks CI rather than shipping.

DavidMetcalfe:
@liuhao1024 — verified this PR's corpus fix against current main (`d188a47e9a`), and it holds up: after merging onto today's main, the `JR_AUTH` hoist lands at the new `skills/web/` path (rename follows the hunk) and `scan_skill(..., source="community")` on the reworded skill returns **verdict `safe`, 0 criticals** under the current scanner. The reword itself is good to go.

**One blocker, in the test you added:** `BUNDLED_SKILL = ... / "skills" / "research" / "blocked-page-recovery"` hardcodes the pre-move path. The skill moved to `skills/web/` (`c49fa88b80`) two hours after this PR's CI last ran, so on a merge with today's main:

```
FAILED tests/tools/test_bundled_skills_self_scan.py::test_bundled_blocked_page_recovery_passes_context_scan
E   FileNotFoundError: ... skills/research/blocked-page-recovery/SKILL.md
1 failed, 1 passed
```

The second test (`..._installable_from_hub`) passes *vacuously* — `scan_skill` on a directory that doesn't exist returns no findings, so `should_allow_install` succeeds without reading any file. Two things fix it:

1. Retarget the constant: `skills/research/...` → `skills/web/...` (one line).
2. Fail loudly on a missing dir — e.g. `assert BUNDLED_SKILL.is_dir()` at the top of each test, or collect `SKILL.md` files and `assert files` first (the existence guard #102476's test uses), so a future path move can never again produce a green run over nothing.

CI status is currently green only because checks last ran before the move — base-branch changes don't retrigger workflows, so the red merge-preview result above hasn't shown up on the PR. The merge itself is conflict-free (`git merge-tree` → 0 conflicts); a rebase + the retarget is the whole delta.

liuhao1024:
@DavidMetcalfe Both fixes applied — rebased onto current main (conflict-free, as your `git merge-tree` showed) plus the two-part test hardening:

1. `BUNDLED_SKILL` retargeted to `skills/web/blocked-page-recovery` — the rebase carried the SKILL.md hunk to the new path automatically, only the test constant needed the one-line fix.
2. Existence guard in **both** tests: `assert BUNDLED_SKILL.is_dir(), ...` before any scan, so a future move can never again produce a green run over nothing — I verified the guard goes red against the old `skills/research/` path while the `scan_skill`-on-missing-dir vacuous pass is exactly what it prevents.

`pytest tests/tools/test_bundled_skills_self_scan.py` → **2 passed** (non-vacuous now). Pushed; since base-branch changes don't retrigger workflows, this push should also surface the first post-move CI run on the branch.

DavidMetcalfe:
@liuhao1024 — verified at the current head (`740e8cd`): `BUNDLED_SKILL` now targets `skills/web/blocked-page-recovery`, and both tests open with the `is_dir()` existence guard. The blocker is cleared.

`tests/tools/test_bundled_skills_self_scan.py` passes (2 tests) and fails closed on credential-interpolating one-liners (`['exfil_curl']`, `allowed=False`) and on an invalid skill directory path. No commit on `main` since the merge-base (`7de8728c`) touches the skill, these tests, or the two scanner modules, so the result holds against today's main.

One leftover: the PR body still cites the old `skills/research/blocked-page-recovery/SKILL.md` path in "Changes Made" and in the "How to Test" snippet — worth updating to `skills/web/` so the instructions match the tree.

Note on CI: the runs for the new head exist but are parked at `action_required` (awaiting approval), so the first post-move run hasn't executed yet.

liuhao1024:
@DavidMetcalfe Good catch on the leftover — PR body updated: both the "Changes Made" entry and the "How to Test" snippet now reference `skills/web/blocked-page-recovery` (2 occurrences, matching the tree). Thanks for verifying the existence-guard fix and the fail-closed behavior.

PR #98490, opened 2026-08-30
fix(skills): keep bundled blocked-page recovery available

### LINKED ISSUE #98474: Bundled skill research/blocked-page-recovery trips the exfil_curl context scanner — silently dropped from the prompt on every fresh v0.20.6 install

fangliquanflq:
## What does this PR do?

The bundled `research/blocked-page-recovery` skill is currently replaced with a blocked placeholder before it can reach the system prompt, so fresh installations silently lose the recovery workflow. This change separates the Jina authorization value from the `curl` invocation and adds a regression test that runs the shipped skill through the same context scanner used by the prompt builder.

### Symptom

Loading the bundled skill causes `scan_for_threats(..., scope="context")` to return `exfil_curl`, and the prompt builder replaces the complete skill with a blocked placeholder.

### Impact

Users cannot invoke the bundled blocked-page recovery instructions even though the skill is installed.

### Bug Cause

**Trigger:** `skills/research/blocked-page-recovery/SKILL.md:97` puts a `curl` command and a secret-suffixed environment variable on the same line.

**Causal chain:**
1. The bundled skill text is scanned before prompt construction.
2. The same-line Jina command matches the `exfil_curl` context pattern.
3. The prompt builder blocks the complete skill instead of loading its instructions.

**Why it is wrong:** The command is a legitimate authenticated request to the documented Jina endpoint, but its example has the same command shape as a secret exfiltration payload.

**Working sibling / contrast:** Curl examples without a secret-suffixed variable on the command line pass the scanner. The security pattern remains unchanged and continues to catch real exfiltration shapes.

**Ruled out:** The current `main` branch still returns `['exfil_curl']` for this complete skill, so the recent suffix-boundary scanner fix does not cover this case.

### Fix

Build the authorization header in a separate shell variable, then pass that non-secret-suffixed variable to `curl`. A focused regression test asserts that the shipped skill remains accepted by the context scanner.

## Related Issue

Closes #98474

## Type of Change

## Changes Made

- `skills/research/blocked-page-recovery/SKILL.md` - split the Jina authorization header from the curl invocation.
- `tests/tools/test_threat_patterns.py` - verify the bundled skill passes the runtime context scan.

## How to Test

1. Run the threat-pattern test file.
2. Confirm all 29 tests pass, including the bundled skill regression.

```bash
scripts/run_tests.sh tests/tools/test_threat_patterns.py -q
```

### Code

- [x] I've run the relevant repository test entry and all targeted tests pass

- [x] I've updated relevant documentation - the affected bundled skill is the documentation surface
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) - the shell command remains portable across the skill's declared platforms

### DIFF
--- a/skills/research/blocked-page-recovery/SKILL.md
+++ b/skills/research/blocked-page-recovery/SKILL.md
@@ -94,7 +94,8 @@ returns markdown. Anonymous access is dead (401 → Turnstile); a key is
 required:
 
 ```bash
-curl -s -H "Authorization: Bearer $JINA_API_KEY" "https://r.jina.ai/{URL}"
+JR_AUTH="Authorization: Bearer $JINA_API_KEY"
+curl -s -H "$JR_AUTH" "https://r.jina.ai/{URL}"
 ```
 
 Handles JS SPAs that archives can't. Skip this route entirely when the env
--- a/tests/tools/test_threat_patterns.py
+++ b/tests/tools/test_threat_patterns.py
@@ -6,6 +6,7 @@ helpers `scan_for_threats()` / `first_threat_message()`.
 """
 
 import time
+from pathlib import Path
 
 import pytest
 
@@ -33,6 +34,21 @@ class TestScopes:
         text = "ignore previous instructions"
         assert "prompt_injection" in scan_for_threats(text, scope="all")
         assert "prompt_injection" in scan_for_threats(text, scope="strict")
+
+
+class TestBundledSkillSafety:
+    def test_blocked_page_recovery_passes_context_scan(self):
+        skill_path = (
+            Path(__file__).resolve().parents[2]
+            / "skills"
+            / "research"
+            / "blocked-page-recovery"
+            / "SKILL.md"
+        )
+
+        assert scan_for_threats(
+            skill_path.read_text(encoding="utf-8"), scope="context"
+        ) == []
 
 
 # =========================================================================


alt-glitch:

Duplicate of #98489: both rewrite the bundled Jina Reader curl example to avoid the same exfil_curl context-scan false positive. #98489 was opened first and also tests the Hub/Skills Guard path.

fangliquanflq:
Confirmed. PR #98489 was opened 24 seconds earlier and applies the identical `SKILL.md` production hunk: it assigns the Jina authorization header to `JR_AUTH` before passing that variable to `curl`. It also covers both affected paths with fail-closed tests: the bundled context scan (`scan_for_threats(..., scope="context")`) and Hub/Skills Guard installation (`scan_skill` / `should_allow_install`). This PR covers the first path only, and its current required CI checks pass, so there is no distinct code change to make here.

DavidMetcalfe:
@fangliquanflq — validated this PR's corpus fix on current main (`d188a47e9a`): the `JR_AUTH` hoist merges cleanly (rename follows the hunk to `skills/web/blocked-page-recovery/`) and the reworded skill scans **`safe`, 0 criticals** under the current scanner. The reword is solid.

Same test-path blocker as #98489, though: the new `TestBundledSkillSafety` test hardcodes `skills/research/blocked-page-recovery`, which no longer exists after the `skills/web/` move — on a merge with today's main it fails `FileNotFoundError`. Retarget to `skills/web/...` and consider an `assert skill_path.is_dir()` first so a future move fails with an explicit assertion rather than an unhandled read error.

One coordination note since this matters for #98474 / #111334 class 2: your `SKILL.md` hunk is textually identical to the one in #98489 (same hoist, same lines) — both open PRs carry the same corpus fix. #98489 pairs it with a `should_allow_install` (install-verdict) assertion, which is the check #98474 is actually about, while yours asserts the context-scope scan only. Two near-identical PRs split the review load for the same one-line fix, so consolidating on whichever vehicle lands first is probably less work for everyone — which one that is stays your call; nothing here asks you to close anything.

Whichever direction you take, the retarget above is required either way.

fangliquanflq:
Thanks for checking. I merged current main and retargeted the regression test to skills/web/blocked-page-recovery, with an explicit directory assertion before reading SKILL.md (349609e7b657). The JR_AUTH change survived the rename. Verification via scripts/run_tests.sh: test_threat_patterns.py 58 passed; test_skills_guard.py 56 passed; test_authoring_standards.py 1,457 passed. The adjacent context_file_sources test has one unrelated Windows path-separator failure (expected ../AGENTS.md, got ..\AGENTS.md) in untouched code. Agreed that #98489 has the same skill hunk and additionally checks the Hub install verdict; this PR still covers only the context-scope path. I am leaving consolidation to the maintainers rather than closing either PR on this notification.

DavidMetcalfe:
@fangliquanflq — verified at 349609e7b6: the retarget to `skills/web/blocked-page-recovery` and the `skill_path.parent.is_dir()` assertion resolve both points I raised.

On this head: `tests/tools/test_threat_patterns.py` 58 passed, `tests/tools/test_skills_guard.py` 56 passed, `tests/skills/test_authoring_standards.py` 1,457 passed — your counts match on macOS. The new `TestBundledSkillSafety` test fails if the `JR_AUTH` hoist is reverted and passes with it in place, so it pins the fix. `tests/agent/test_context_file_sources.py` also passes here (2 passed) and isn't touched by this diff, consistent with the Windows-only diagnosis.

Nothing further needed from me on this PR.

fangliquanflq:
Thanks for verifying the updated head. The regression test now targets `skills/web/blocked-page-recovery`, asserts the directory exists, and covers the context scan; the `JR_AUTH` hoist remains in the skill diff. Your macOS runs passed `test_threat_patterns.py` (58), `test_skills_guard.py` (56), `test_authoring_standards.py` (1,457), and `test_context_file_sources.py` (2). No further change is needed for this feedback.

PR #92273, opened 2026-08-22
fix(bedrock): add claude-opus-5 to context length table

bagg3rs:
## Problem

`claude-opus-5` is missing from `BEDROCK_CONTEXT_LENGTHS` in `agent/bedrock_adapter.py`, so `get_bedrock_context_length()` falls through to `BEDROCK_DEFAULT_CONTEXT_LENGTH` (128K) instead of the model's actual 1M window.

Because the context compressor's threshold defaults to 50% of the resolved context length, any Bedrock user running Opus 5 (`anthropic.claude-opus-5`, or regional profiles like `eu.anthropic.claude-opus-5`) gets compaction triggering at just **64K tokens**. In practice an agentic session with a few file reads and tool calls hits "🗜️ Compacting context" every few minutes, and long sessions degrade through repeated re-summarization ("Session compressed 9 times — accuracy may degrade").

The rest of the Claude 5 family (`claude-fable-5`, `claude-sonnet-5`) is already in the table at 1M — Opus 5 just missed the update.

## Fix

One table entry:

```python
"anthropic.claude-opus-5":       1_000_000,
```

Substring matching handles the regional inference-profile prefixes (`eu.`/`us.`/`global.`) and versioned IDs, same as the existing entries. The `anthropic.claude-opus-4` fallback entry is unaffected (longest-key-wins).

## Testing

Added a regression test asserting 1M for the bare ID and `eu.`/`us.` profile variants. `tests/agent/test_bedrock_adapter.py`: 67 passed.

Found in the wild: my gateway on Bedrock eu Sonnet 5/Opus 5 was compacting constantly until I traced it to this lookup.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #31277, opened 2026-05-24
[Bug]: Native Bedrock Converse adapter does not forward context-1m beta — Opus 4.6/4.7 stuck at 200K (PR #16793 only patched OpenAI-compat path)


pgregg88:
## Summary

The native Bedrock Converse adapter (`agent/bedrock_adapter.py`) does not forward the `context-1m-2025-08-07` beta header to `additionalModelRequestFields`, which blocks 1M-context for Claude Opus 4.6 / 4.7 on accounts with the entitlement.

PR #16793 (merged 2026-04-28) fixed this for the OpenAI-compat path through the `AnthropicBedrock` client (`agent/anthropic_adapter.py` — `_CONTEXT_1M_BETA = "context-1m-2025-08-07"` machinery on lines 267, 558, 792). The native Converse path was missed.

Users on the native `bedrock_converse` provider (which the setup wizard recommends for direct AWS Bedrock without a Mantle/proxy) are stuck at 200K even with the entitlement.

## Reproduction

1. `~/.hermes/config.yaml` configured for native Bedrock:
   ```yaml
   provider: bedrock
   model: us.anthropic.claude-opus-4-7
   ```
   (No `OPENAI_BASE_URL`/Mantle proxy — direct boto3 path.)

2. Confirm AWS account has the 1M context beta entitlement (Anthropic gate it; was on the allow-list for me).

3. Start Hermes, observe banner reports 200K context for `claude-opus-4-7`.

4. Inspect `build_converse_kwargs()` in `agent/bedrock_adapter.py` — `additionalModelRequestFields` is never populated with `anthropic_beta`.

5. `BEDROCK_CONTEXT_LENGTHS` table doesn't list `claude-opus-4-7` at all; it falls through to the `claude-opus-4` substring match (200K) or the 128K default.

## Expected

When user opts in (their AWS account has the entitlement), `bedrock_converse` adapter sends:
```python
additionalModelRequestFields={"anthropic_beta": ["context-1m-2025-08-07"]}
```
and `get_bedrock_context_length("us.anthropic.claude-opus-4-7")` returns `1_000_000`.

## Verified working

I wrote a local patch that:

1. Adds opt-in via `HERMES_BEDROCK_1M_CONTEXT` env var (default OFF, so accounts without the entitlement are unaffected).
[...]

PR #83071, opened 2026-08-10
fix(bedrock): claude-opus-5 missing from BEDROCK_CONTEXT_LENGTHS

nickkpoon:
claude-opus-5 was added to `DEFAULT_CONTEXT_LENGTHS` in 306c9f7661, but `BEDROCK_CONTEXT_LENGTHS` was not updated — even though its own comment requires the 1M entries to match:

```
# These 1M entries must match agent/model_metadata.py
# DEFAULT_CONTEXT_LENGTHS or the agent compresses context prematurely.
```

Bedrock ids are resolved by longest-substring match, so `anthropic.claude-opus-5` had no entry, fell through to the generic `anthropic.claude-opus-4`, and resolved to **200K**.

### Effect

On Bedrock the agent compresses at 200K for a model that serves 1M, silently discarding context the user already paid to load. No error surfaces — it just quietly truncates.

### Verification

The 1M window is real; a live Anthropic endpoint rejects an oversized prompt with its actual ceiling:

```
prompt is too long: 8460056 tokens > 1000000 maximum
```

### Tests

Rather than hardcode another model list that can drift the same way, the added test derives the expected value from `DEFAULT_CONTEXT_LENGTHS`, so a future 1M model only has to be added in one place for a mismatch to fail CI. Confirmed it fails without the one-line fix:

```
anthropic.claude-opus-5: bedrock=None metadata=1000000
```

A second test guards the opposite direction — `haiku-4-5`, `sonnet-4-5`, and `opus-4` must stay 200K, so a longest-key regression that over-advertises context (which would cause hard API errors rather than silent truncation) also fails.

`tests/agent/test_bedrock_1m_context.py` — 4 passed.

PR #75824, opened 2026-08-01
fix(bedrock): add Claude Opus 5 to the 1M context table (salvage #75432)

ryrenz:
## Summary

Bedrock now probes the live model when a region is available, but offline and display paths still use `BEDROCK_CONTEXT_LENGTHS`. That static fallback has no Opus 5 entry, so bare, regional, global, and versioned Opus 5 IDs resolve to 128K instead of their 1M window.

This is the remaining Opus 5 half of #74263; the Sonnet 5 side already landed via #67977.

Fixes #74263

## Root cause

`agent/model_metadata.py::DEFAULT_CONTEXT_LENGTHS` declares `claude-opus-5` as 1M, while the Bedrock fallback table has no matching key. With probing disabled or unavailable, longest-substring lookup finds no match and returns `BEDROCK_DEFAULT_CONTEXT_LENGTH` (128K).

## Fix

- Add `anthropic.claude-opus-5: 1_000_000` to the Bedrock static fallback, sourced from the AWS model card.
- Exercise the production resolver with `probe=False` across bare, `us.`/`eu.`/`au.`/`global.` and version-suffixed IDs.
- Add a behavioral invariant that keeps 1M Claude entries aligned with `DEFAULT_CONTEXT_LENGTHS` without freezing a catalog snapshot.

Salvages #75432, whose author closed it as housekeeping and explicitly invited the work to be carried forward.

## Source

AWS documents Claude Opus 5 on Bedrock with a 1M context window:
https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-5.html

## Validation

- `scripts/run_tests.sh tests/agent/test_bedrock_adapter.py -q`: 69 passed, 5 skipped
- Bedrock/model-metadata call chain (6 files): 218 passed, 5 skipped
- Reverse check: removing the new key fails all 8 regression cases at the 128K fallback
- `ruff check`: passed
- Windows footgun check: passed
- `git diff --check`: passed

PR #98494, opened 2026-08-30
fix(bedrock): add the missing claude-opus-5 context row, and guard the static table against silent drift

JiaDe-Wu:
> **Update (2026-08-30):** this PR now carries two commits. It started as tests-only; live measurement then turned up a real bug that the tests were blind to, so the fix for it is here too.
>
> | commit | what |
> |---|---|
> | `c23c56b` | the guard tests described below (tests-only) |
> | `9432837` | adds the missing `anthropic.claude-opus-5` row **plus** the reverse-direction guard that catches missing rows |
>
> The second commit's one-line table fix is the same one @bagg3rs proposed in #92273 and is credited to them with a `Co-authored-by:` trailer. **If #92273 merges first, I'll rebase this down to the test alone.** I carry it here only because the new guard fails without it — which is the entire point of the guard.

## The bug the first version of this PR could not see

Sweeping all 13 `us.anthropic.*` inference profiles in `us-east-2` against the window Bedrock actually enforces, there is exactly one divergence:

| inference profile | static table | Bedrock enforces | `x-amzn-RequestId` |
|---|---|---|---|
| `us.anthropic.claude-opus-5` | **128,000** | **1,000,000** | `bde3ddfa-1898-49f8-808d-3d9cd326b2c6` |

`claude-opus-5` is in `model_metadata.DEFAULT_CONTEXT_LENGTHS` at 1M but had no `BEDROCK_CONTEXT_LENGTHS` row, so it fell to `BEDROCK_DEFAULT_CONTEXT_LENGTH` (128K) and compression fired at 64K. The symptom is constant "🗜️ Compacting context" with nothing naming a context window — which is why it went unnoticed.

The cross-file check in the first commit **only compares rows the Bedrock table already has**, so it could not see this: the row was absent, not wrong. That is a genuine gap in my own test. The new `test_every_claude_model_in_metadata_is_covered` walks the other direction and reports precisely this on unpatched `main`:

```
claude-opus-5: metadata 1,000,000 vs Bedrock 128,000 (no row at all — falls to the default)
```

All 8 `claude-*` slugs in `DEFAULT_CONTEXT_LENGTHS` are checked. Anthropic sometimes ships to their own API before Bedrock, so a slug genuinely absent from Bedrock goes in `_METADATA_SLUGS_NOT_ON_BEDROCK` with a note rather than having a Bedrock row invented for it; that set is empty today.

I also expected `claude-opus-4-5` to be a second missing row. Measured live it is 200,000 and the generic `anthropic.claude-opus-4` fallback already gives 200,000 — so that guess was wrong and there is no second gap.

Offline suite is now 72 passed (was 69). Everything below still applies.

---

## What

Tests only. Adds `tests/agent/test_bedrock_context_window_guard.py` (+258) — nothing under `agent/` changes.

`BEDROCK_CONTEXT_LENGTHS` is the fallback `get_bedrock_context_length()` uses whenever the live probe can't run: `probe=False`, no region, no credentials, offline. Two invariants are already written down in that code as prose, and neither had a test:

- `get_bedrock_context_length()`'s docstring: *"a stale entry silently caps the window (e.g. a 1M-token Opus pinned to 200K via an `opus-4` substring match)"*
- the table's own comment: *"These 1M entries must match `agent/model_metadata.py` `DEFAULT_CONTEXT_LENGTHS` or the agent compresses context prematurely"*

Both failure modes are silent. Nothing raises; the CLI just advertises the wrong window and compression fires at the wrong threshold.

## The three offline properties

All hold on `main` today — this PR is a guard, not a fix.

**1. Every row is reachable.** Matching is `key in model_id.lower()`, so a key carrying an uppercase character can never match any input; the model silently falls to a shorter row. Nothing else in the suite would surface that typo.

**2. Removing a row can only lower the result.** This is the state of *every* model AWS ships before this table is updated, so it's the common case, not the exotic one. The two directions are not symmetric:

- falling **low** → premature context compression, wasteful but safe;
- falling **high** → the agent packs a prompt the model rejects outright.

So no generic row may be more generous than the specific rows it covers. Concretely, dropping `claude-opus-4-8` must land on `claude-opus-4` (200K), never above 1M. Longest-substring means a *shorter* generic key can never outrank a longer specific one, so this fallback direction is the only one worth asserting — an earlier draft of this test asserted "no row is shadowed", which turns out to be vacuously true and is not in this PR.

**3. The Claude rows agree with `DEFAULT_CONTEXT_LENGTHS`** — the cross-file invariant the table comment requires. 7 rows have a counterpart and all 7 match.

Plus current-generation resolution cases: `us.`/`global.`/`eu.`/`apac.` inference-profile prefixes and `-v1` / `-v1:0` / `-20250514-v1:0` suffixes must not change the answer, and `us.anthropic.claude-opus-4-1-v1:0` must stay 200K rather than be captured by a 1M row.

69 tests, 1.5s, no network.

## Each property was mutation-tested

A guard that can't fail isn't a guard. Every assertion was verified to break under injected drift:

| Injected defect | Result |
|---|---|
| drop the `anthropic.claude-opus-4-8` row | `us.anthropic.claude-opus-4-8` → 200,000; `test_versioned_claude_ids_beat_the_generic_opus_4_row` fails |
| add `anthropic.claude-Opus-4-9` (uppercase) | `test_every_row_is_reachable[anthropic.claude-Opus-4-9]` fails |
| add `anthropic.claude-sonnet: 2_000_000` | `test_removing_a_row_can_only_lower_the_window` fails for `claude-sonnet-4` **and** `claude-sonnet-5` |
| set `anthropic.claude-sonnet-5: 200_000` | `test_claude_rows_match_default_context_lengths` fails |
| add `anthropic.claude-opus-4-1: 1_000_000` | `test_opus_4_1_is_not_over_matched_to_1m` fails |

That third one is how I found the vacuous assertion: injecting a shadowing row produced **zero** failures, because it can't shadow. I replaced it rather than keep a test that always passes.

## Opt-in live check

`TestTableMatchesLiveBedrock` asks the question the offline tests can't: does the table still match what Bedrock actually enforces? It **reuses `probe_bedrock_context_length()`** instead of reimplementing the probe, so it exercises production code at the same time.

Gated twice — `@pytest.mark.integration` (excluded by the default `-m 'not integration'`) plus `HERMES_E2E_BEDROCK=1`, following the `tests/tools/test_browser_supervisor.py` precedent. It costs **zero tokens**: length validation runs before inference, so an oversized prompt is rejected immediately and the error reports the real `maximum`.

```
HERMES_E2E_BEDROCK=1 AWS_REGION=us-east-2 \
    pytest tests/agent/test_bedrock_context_window_guard.py -m integration -v
```

Verified against live Bedrock in `us-east-2` on 2026-08-30 — 5 passed in 7.6s:

| model | static table | Bedrock reports | `x-amzn-RequestId` |
|---|---|---|---|
| `us.anthropic.claude-sonnet-4-6` | 1,000,000 | 1,000,000 | `a7a337da-3afb-4c3c-9f16-b5dede06395a` |
| `us.anthropic.claude-opus-4-6-v1` | 1,000,000 | 1,000,000 | `8c8a77ae-21a0-42b5-8e1d-e484f5ea9322` |
| `us.anthropic.claude-opus-4-7` | 1,000,000 | 1,000,000 | `dca8cb34-f8a3-4d44-ac4c-ebba133d8d1a` |
| `us.anthropic.claude-opus-4-8` | 1,000,000 | 1,000,000 | `e0993efd-519a-4ce4-9bf6-2d1d2bf62232` |
| `us.anthropic.claude-sonnet-5` | 1,000,000 | 1,000,000 | `d37e8a19-b8a2-44ce-8175-ac54819dcab4` |

Default run (no env var) correctly skips all 5 with `live Bedrock: set HERMES_E2E_BEDROCK=1 to opt in`.

## One thing worth a maintainer's attention

The live group did not work on the first attempt, and the reason generalises beyond this PR.

The autouse isolation fixture in `tests/conftest.py` deletes `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN` and sets `AWS_EC2_METADATA_DISABLED=true`, commented *"Tests don't run on EC2 — IMDS is always unreachable here."* That default is correct and I am not proposing to change it — it is what stops a unit test touching a real account.

But its effect on an **opt-in** live test is a false negative, not a failure: signing fails instantly, `probe_bedrock_context_length()` returns `None` per its contract, and the test **skips itself with a misleading "no credentials" reason**. My first run was `5 skipped in 0.94s` — fast enough that it clearly never called Bedrock, which is the only reason I noticed rather than believing the skip.

The fix stays inside this test file: the `AWS_*` environment is snapshotted at **collection** time — before the autouse fixture runs, the only point where the real values are still visible — and restored for the duration of the live test only, with `reset_client_cache()` called on both entry and exit so a live-credentialled client can't leak into a unit test.

Any future integration test needing real AWS will hit the same wall. If maintainers would rather have a shared fixture for this in `tests/conftest.py`, I'm happy to follow up.

## Context

Came out of measuring #31277 (Bedrock 1M context). That issue's remaining half turns out to be obsolete — the `context-1m-2025-08-07` beta was retired on 2026-04-30 when 1M went GA, and all five current Claude models now report 1,000,000 with *and* without the beta header, so gating 1M behind an opt-in would now re-introduce the under-reporting the issue was filed about. Details and measurements in [my comment there](https://github.com/NousResearch/hermes-agent/issues/31277#issuecomment-5467962086). This PR is the artifact I offered in that comment instead: the table is right today, so lock in that it stays right.

## Testing

- `pytest tests/agent/test_bedrock_context_window_guard.py` → 69 passed, 5 deselected, 1.5s
- `pytest tests/agent/test_bedrock_adapter.py tests/agent/test_model_metadata.py tests/agent/test_bedrock_1m_context.py tests/agent/test_bedrock_empty_text_blocks.py tests/agent/test_bedrock_context_window_guard.py` → 261 passed
- `ruff check .` → clean
- live group on real Bedrock → 5 passed (table above)

### DIFF
--- a/agent/bedrock_adapter.py
+++ b/agent/bedrock_adapter.py
@@ -1782,6 +1782,7 @@ BEDROCK_CONTEXT_LENGTHS: Dict[str, int] = {
     "anthropic.claude-fable-5":      1_000_000,
     "anthropic.claude-fable":        1_000_000,
     "anthropic.claude-sonnet-5":     1_000_000,
+    "anthropic.claude-opus-5":       1_000_000,
     "anthropic.claude-opus-4-8":     1_000_000,
     "anthropic.claude-opus-4-7":     1_000_000,
     "anthropic.claude-opus-4-6":     1_000_000,
--- a/tests/agent/test_bedrock_context_window_guard.py
+++ b/tests/agent/test_bedrock_context_window_guard.py
@@ -64,6 +64,14 @@ _AWS_ENV_AT_COLLECTION = {
     k: v for k, v in os.environ.items() if k.startswith("AWS_")
 }
 
+# Escape hatch for the coverage check below: Claude slugs that model_metadata
+# knows but Bedrock genuinely does not offer, so BEDROCK_CONTEXT_LENGTHS is right
+# to stay silent about them. Empty today — every claude-* slug in
+# DEFAULT_CONTEXT_LENGTHS is served by Bedrock. Anthropic sometimes ships to
+# their own API before Bedrock, so add a slug here (with a note) rather than
+# inventing a Bedrock row for a model that isn't there.
+_METADATA_SLUGS_NOT_ON_BEDROCK: frozenset = frozenset()
+
 # Measured against live Bedrock, us-east-2, 2026-08-30. See module docstring.
 LIVE_VERIFIED_WINDOWS = {
     "us.anthropic.claude-sonnet-4-6": 1_000_000,
@@ -184,6 +192,50 @@ class TestTableAgreesWithModelMetadata:
             + "\n  ".join(divergences)
         )
         assert compared, "no rows compared — the key-slug mapping has drifted"
+
+    def test_every_claude_model_in_metadata_is_covered(self):
+        """The other direction: a *missing* row, which is the one that bites.
+
+        The test above only compares rows the Bedrock table already has, so it
+        is blind to the failure that actually happens — a model the rest of the
+        codebase knows about having no Bedrock row at all, and silently
+        resolving to the 128K default. That halves the compression threshold to
+        64K, and the symptom users report is constant "🗜️ Compacting context"
+        rather than anything naming a context window.
+
+        This is exactly how ``claude-opus-5`` was missed (#92273): present in
+        DEFAULT_CONTEXT_LENGTHS at 1M, absent from BEDROCK_CONTEXT_LENGTHS.
+        """
+        from agent.model_metadata import DEFAULT_CONTEXT_LENGTHS
+
+        checked = 0
+        gaps = []
+        for slug, want in sorted(DEFAULT_CONTEXT_LENGTHS.items()):
+            if not slug.startswith("claude-"):
+                continue
+            if "." in slug:
+                continue  # dotted aliases (claude-opus-4.6) are not Bedrock IDs
+            if slug in _METADATA_SLUGS_NOT_ON_BEDROCK:
+                continue
+            checked += 1
+            got = _static_bedrock_context_length(f"anthropic.{slug}")
+            if got != want:
+                how = (
+                    "no row at all — falls to the default"
+                    if got == BEDROCK_DEFAULT_CONTEXT_LENGTH
+                    else "resolves via a different row"
+                )
+                gaps.append(f"{slug}: metadata {want:,} vs Bedrock {got:,} ({how})")
+
+        assert not gaps, (
+            "Claude models known to model_metadata do not resolve to the same "
+            "window on Bedrock:\n  "
+            + "\n  ".join(gaps)
+            + "\n\nAdd the missing row to BEDROCK_CONTEXT_LENGTHS. If the model "
+            "genuinely is not offered on Bedrock, add its slug to "
+            "_METADATA_SLUGS_NOT_ON_BEDROCK in this file with a note saying so."
+        )
+        assert checked, "no slugs checked — the claude- prefix filter has drifted"
 
 
 @pytest.mark.integration
--- a/tests/agent/test_bedrock_context_window_guard.py
+++ b/tests/agent/test_bedrock_context_window_guard.py
@@ -1,0 +1,258 @@
+"""Guards on the Bedrock static context-window fallback table.
+
+``get_bedrock_context_length()`` prefers a live probe and falls back to
+``BEDROCK_CONTEXT_LENGTHS``. Its own docstring names the failure mode the
+fallback is prone to:
+
+    a stale entry silently caps the window (e.g. a 1M-token Opus pinned to
+    200K via an ``opus-4`` substring match)
+
+That is a *silent* failure: nothing raises, the CLI just advertises the wrong
+window and context compression fires several times earlier than it needs to.
+The offline/display paths (``probe=False``, or no region) have only the table,
+so drift there is not masked by the probe.
+
+Three properties are asserted offline, all of which the table currently
+satisfies and none of which was covered:
+
+* **Every row is reachable.** Matching is ``key in model_id.lower()``, so a key
+  carrying an uppercase character can never match anything and the model
+  silently drops to a shorter row.
+* **Removing a row can only lower the result.** That is the situation for every
+  model AWS ships before this table is updated. Falling low costs premature
+  compression; falling *high* means the agent builds a prompt the model rejects.
+  So no generic row may be more generous than the specific rows it covers.
+* **The Claude rows agree with ``model_metadata.DEFAULT_CONTEXT_LENGTHS``.** The
+  table's own comment requires this ("These 1M entries must match
+  agent/model_metadata.py DEFAULT_CONTEXT_LENGTHS or the agent compresses
+  context prematurely") but nothing enforced it.
+
+A third check compares the table against what Bedrock actually enforces. It
+needs live AWS, so it is opt-in twice over — ``@pytest.mark.integration``
+(excluded by the default ``-m 'not integration'``) plus ``HERMES_E2E_BEDROCK=1``
+— and it reuses the production probe rather than reimplementing it, so it
+exercises ``probe_bedrock_context_length()`` at the same time.
+
+The expected windows below were measured against live Bedrock in ``us-east-2``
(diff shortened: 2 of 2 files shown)


Enough1122:

### PR #98494 — fix(bedrock): add the missing claude-opus-5 context row, and guard the static table against silent drift

**Summary:** Adds missing `anthropic.claude-opus-5: 1_000_000` to `BEDROCK_CONTEXT_LENGTHS` (`agent/bedrock_adapter.py:1785`) — same one-line fix as #9227 — and introduces `tests/agent/test_bedrock_context_window_guard.py` (310 lines, 3 classes + live probe) that fails closed on (1) unreachable rows (key not `in model_id.lower()`), (2) generic-row over-generosity (`fallback <= own` when row removed), (3) exact parity with `model_metadata.DEFAULT_CONTEXT_LENGTHS`, and (4) full reverse coverage (every `claude-*` slug in metadata resolves to same window via `_static_bedrock_context_length`, catching missing rows like opus-5 falling to 128K default and halving threshold to 64K). Live `probe_bedrock_context_length` opt-in (`@pytest.mark.integration` + `HERMES_E2E_BEDROCK=1`) reuses collection-time AWS env capture to survive `tests/conftest.py` AWS scrub.

- `_METADATA_SLUGS_NOT_ON_BEDROCK` escape hatch is empty today (every `claude-*` slug is on Bedrock); comment directs adding a slug there rather than inventing a row for a not-yet-available model.
- Guard correctly excludes dotted aliases (`claude-opus-4.6`) and non-`anthropic.` rows from metadata comparison; `compared`/`checked` asserts prevent silently empty parametrization if filters drift.

**Verdict:** One-line correctness fix plus the strongest static-table guard in the repo. No issues.

JiaDe-Wu:
Closing this — and **not** in favour of a PR of mine. It should have been closed in favour of someone else's, and I should have found that before opening it.

I ran a proper duplicate search today, which I clearly did not do properly in August. Three PRs were already open on exactly this row when I filed this one:

- #75824 (@ryrenz) — oldest, 2026-08-01
- #83071 (@nickkpoon)
- #92273 (@bagg3rs)

**#75824 is the one I'd back.** It is written against the *current* compacted `dict.fromkeys` shape of the table, not the older one-key-per-line shape this PR assumes, so it applies today without a rewrite. And its drift guard is better than the one I wrote here: mine excluded the dotted `DEFAULT_CONTEXT_LENGTHS` spellings (`claude-opus-4.8`) as out of scope, while #75824 normalises dots to hyphens and checks them, which closes the exact hole I left open — a future 1M model that only ever gets a dotted alias would have slipped past my version.

For whatever it is worth to whoever reviews #75824, I verified the defect independently on `7dc796463d` before writing this. Importing the module and calling `get_bedrock_context_length(model_id, probe=False)`, no network:

| model id | resolves to | matched key |
|---|---|---|
| `us.anthropic.claude-opus-5-20260101-v1:0` | **128,000** | **none at all → `BEDROCK_DEFAULT_CONTEXT_LENGTH`** |
| `us.anthropic.claude-sonnet-5-20260101-v1:0` | 1,000,000 | `anthropic.claude-sonnet-5` |
| `us.anthropic.claude-opus-4-8-20260101-v1:0` | 1,000,000 | `anthropic.claude-opus-4-8` |

Worth stressing that it matches **no** key — not even the generic `anthropic.claude-opus-4` — so it does not get quietly pinned to 200K, it falls all the way to the 128K default while `DEFAULT_CONTEXT_LENGTHS["claude-opus-5"]` is 1,000,000. That is the file's own comment at `bedrock_adapter.py:935-936` being violated.

I also audited the rest of the table against `DEFAULT_CONTEXT_LENGTHS` while I was in there, in case it mattered to the guard's scope: every other 1M `anthropic.` row agrees, the nine ids that look absent are all 200K models already covered by the `"claude": 200000` catch-all on the other side, and the dotted spellings are the OpenRouter route's keys. `claude-opus-5` is the only genuine mismatch — so a guard over this invariant should go green immediately once the row lands, with nothing else to clean up first.

Separately, the Llama 4 half of my table work — which nobody else has open — is now #106492.

PR #68341, opened 2026-07-21
feat(prompt): add machine hostname to environment hints

jedi95:
## Problem

`build_environment_hints()` emits OS and kernel release (e.g. `Host: Linux (7.0.0-28-generic)`) but not the machine's network hostname. On multi-host setups the agent has no way to know which host it's already running on, causing it to SSH to itself instead of executing commands directly.

This is a frequent pain point in homelab/multi-server setups where the agent manages several hosts by name — it sees a hostname in instructions, doesn't recognize it as itself, and opens a redundant SSH session to the box it's already on.

## Solution

Add a `Hostname:` line to the local-backend environment hints block, using `platform.node()` with the FQDN suffix stripped (split on first `.`):

```
Host: Linux (7.0.0-28-generic)
Hostname: myhost
User home directory: /home/user
Current working directory: /home/user/code
```

Remote/sandbox backends (docker, ssh, modal) continue to suppress all host info — the agent's tools operate inside the backend, not on the host.

## Edge cases handled

- **FQDN**: `platform.node()` may return `host.example.com`; split on `.` and take the first segment
- **Empty hostname**: if `platform.node()` returns empty string, no `Hostname:` line is emitted (no point showing a blank)
- **Failure**: wrapped in try/except so a `platform.node()` failure can't break prompt building
- **Windows**: the existing Windows-specific note ("hostname is NOT the username") remains intact and is not confused with the new `Hostname:` line

## Tests

- Updated `test_build_environment_hints_on_linux_local` — asserts `Hostname: testhost` present
- Updated `test_build_environment_hints_on_windows_local` — asserts `Hostname: DESKTOP-WIN10` present
- Updated `test_build_environment_hints_on_macos_local` — asserts `Hostname: MacBook-Pro` present
- Updated `test_build_environment_hints_suppresses_host_on_docker_backend` — asserts `Hostname:` absent (remote backend)
- Added `test_build_environment_hints_hostname_strips_fqdn` — verifies FQDN stripping
- Added `test_build_environment_hints_hostname_empty_when_node_returns_empty` — verifies empty hostname is omitted

All 10 environment_hints tests pass. Full prompt_builder + system_prompt + platform_hint test suite (200 tests) passes with no regressions.

## Prefix-cache impact

The hostname is stable for the lifetime of the machine (doesn't change between sessions), so this addition to the stable tier does not invalidate prefix caching. It adds ~5 tokens to the system prompt.

PR #38010, opened 2026-06-03
feat(prompt): surface profile isolation in environment hints

Macgrady4Ever:
## What

When Hermes runs under a non-default profile, `build_environment_hints()` now
adds a short block telling the agent it is profile-scoped: the profile name, its
`HERMES_HOME`, the shared Hermes root, and a warning not to override
`HOME` / `HERMES_HOME` at runtime. For the default profile the environment block
is unchanged.

Closes #38008.

## Why

`build_environment_hints()` reported the OS-level home and cwd but never
mentioned profile scoping. Under a non-default profile, the active `HERMES_HOME`
is `<root>/profiles/<name>` while shared/cross-profile state lives under the
root — and the agent (or any tool/subprocess resolving paths from `$HOME` / `~`)
could silently read/write the wrong profile and fork shared state.

This mirrors guards that already exist on the *write* side — the cross-profile
write guard in `agent/file_safety.py` and the `HERMES_HOME` fallback warning in
`hermes_constants.get_hermes_home()` (#18594) — and closes the agent-facing
**prompt** gap. It also complements the operator-supplied
`HERMES_ENVIRONMENT_HINT` by deriving the same awareness automatically.

## Before / After

Run as a non-default profile (`HERMES_HOME=~/.hermes/profiles/coder`):

**Before**
```
Host: WSL (Windows Subsystem for Linux)
User home directory: /home/user
Current working directory: /home/user/work
```

**After**
```
Host: WSL (Windows Subsystem for Linux)
User home directory: /home/user
Current working directory: /home/user/work

Hermes profile: coder — you are running under an isolated, per-profile Hermes home, not the shared root.
Profile home (HERMES_HOME): /home/user/.hermes/profiles/coder
Shared Hermes root (all profiles): /home/user/.hermes
Per-profile state (skills, plugins, cron, memories) lives under the profile home; shared or cross-profile resources live under the shared root. Tools and subprocesses may run with a per-profile HOME, so paths built from $HOME / ~ can resolve INTO this profile rather than the shared root. Do NOT override HOME or HERMES_HOME at runtime to work around a path that looks wrong — that masks a real path bug and can silently fork shared state. Use absolute paths under the shared root for shared resources, and report path bugs instead.
```

## Implementation

- New helper `_build_profile_scope_hint()` in `agent/prompt_builder.py`. It
  reuses existing detection — `file_safety._resolve_active_profile_name()`,
  `hermes_constants.get_hermes_home()` / `get_default_hermes_root()`.
[...]

PR #98497, opened 2026-08-30
feat: surface two SSH gaps to the model — shell-out guidance + backend target host

Sheardo:
## What does this PR do?

The SSH backend works. Two things about working with SSH are invisible to the model, and this PR surfaces both. No behavior change to the backend, no new tool, no new config — these are prompt-surface fixes.

**1. The model isn't told how to shell out to `ssh`.** The SSH backend covers "run everything on that host". It doesn't cover "reach a host other than the configured one" — a deploy target, a box named mid-conversation, `scp` to somewhere else. That is a shell-out through `terminal`, and two properties of this environment make it fail in ways nothing on the tool surface hints at:

*Commands run with stdin on `/dev/null`* — `stdin=subprocess.DEVNULL` in `environments/local.py:1540` and `environments/base.py:371`. So ssh cannot prompt; password, passphrase, and host-key confirmation all fail unanswered rather than asking:

```
$ ssh -o StrictHostKeyChecking=accept-new user@host   # via terminal
Pseudo-terminal will not be allocated because stdin is not a terminal.
```

`pty=true` fixes this and has supported the local and SSH backends all along — but its description advertised only "Codex, Claude Code, or Python REPL", so the model never reached for it.

*The only in-tree ssh example teaches `UserKnownHostsFile=/dev/null`* — `optional-skills/devops/pinggy-tunnel/SKILL.md` uses it in eight places, and it's what agents copy because it's the only example they have. Correct for a throwaway tunnel endpoint; wrong for a real host, where it discards the host key on every connection so nothing is ever pinned. Where the known-hosts target isn't writable, ssh additionally emits

```
Failed to add the host to the list of known hosts (/dev/null).
```

which reads as a failure in tool output. That one is a warning, not fatal — the connection proceeds — but it's noise the agent then has to reason about.

**2. On the ssh backend, the model isn't told which host it's on.** `build_environment_hints` correctly suppresses host info for remote backends and probes the backend for OS, `$HOME`, cwd and user. For ssh that leaves a gap the probe can't close: it reports what the remote box *looks like* but never *which box it is*. The agent sees "a remote host reached over SSH" and has no way to name the machine it's standing on — so it can't reason about which host a path, service, or log belongs to. Now:

```
Terminal backend: ssh (ubuntu@build-01.example.com). Your `terminal`, `read_file`, ...
```

### Why this approach

Both additions live in the **stable, cached** prompt tier, read once at build time and static per process — the same mechanism as the backend detection immediately above them. Nothing per-turn, nothing that mutates a cached prefix.

`_ssh_target_summary()` deliberately does **not** scan `~/.ssh/config`: that would put the user's whole host inventory into every prompt, and `~/.ssh` is already treated as strict-sensitive by `tools/approval.py` and `tools/threat_patterns.py`. It emits the destination only — the key path never enters the prompt (there's a test for that). Port 22 is elided as noise.

The `BatchMode=yes` shape of the guidance matches what `hermes_cli/doctor.py:1994` and `hermes_cli/setup.py:1629` already do when they probe an SSH host, so the model is pointed at the same idiom the CLI uses.

## Related Issue

No existing issue — found while debugging why an agent asked to reach a remote box emitted `Failed to add the host to the list of known hosts (/dev/null)` and then failed to get a session, rather than connecting cleanly. Happy to file one if you'd prefer that first.

I searched open and merged PRs/issues for overlap. Closest, neither overlapping: #38010 (profile isolation in environment hints) and #68341 (hostname in environment hints).

## Type of Change

## Changes Made

- `tools/terminal_tool.py` — one line added to `TERMINAL_TOOL_DESCRIPTION` pointing at `-o BatchMode=yes -o StrictHostKeyChecking=accept-new` for one-shot commands and `pty=true` (scoped to the local/ssh backends, the only ones that support it) when a prompt is unavoidable; `ssh` named in the `pty` parameter description.
- `agent/prompt_builder.py` — new `_ssh_target_summary()`; `Terminal backend: ssh (user@host[:port])` in both the probe-success and probe-fallback branches.
- `tests/agent/test_prompt_builder.py` — 5 tests: target on the probe-success and probe-fallback paths, non-default-port-only, empty when host unset, key-path non-leakage (+ host suppression still holds).
- `tests/tools/test_terminal_tool.py` — 2 tests pinning the ssh guidance and the `pty` description.
- `website/docs/user-guide/features/tools.md` — the PTY paragraph described only "Codex and Claude Code" and omitted both ssh and the reason the flag exists.

## How to Test

1. `scripts/run_tests.sh tests/agent/test_prompt_builder.py tests/tools/test_terminal_tool.py`
2. Configure `TERMINAL_ENV=ssh` with `TERMINAL_SSH_HOST` / `TERMINAL_SSH_USER` and start Hermes. The environment hint now names the target instead of the bare `Terminal backend: ssh.`
3. On a local backend, ask the agent to run something on a remote host. It should reach for key auth + `BatchMode`/`accept-new`, or `pty=true`, instead of a bare `ssh` that dies on the `/dev/null` stdin.

### Code

- [ ] I've run `pytest tests/ -q` and all tests pass — **partial, see below**
- [x] I've tested on my platform: Ubuntu on WSL2, Python 3.12

On the test checkbox: I ran `tests/agent` and `tests/tools` rather than the whole suite, in a venv built from the standard install plus `pyyaml`/`requests`/`httpx`/`openai`/`python-dotenv`/`wcwidth`. Six files under `tests/tools` (browser_cdp, clipboard, mcp_oauth_metadata, mcp_tool, skill_bundle_provenance, slack_send_message_media) couldn't be collected in my env for missing optional deps and were skipped. Everything that could run passed. CI is the authoritative check here.

- [x] I've updated relevant documentation — `build_environment_hints` docstring, and the PTY paragraph in `website/docs/user-guide/features/tools.md`
- [x] `cli-config.yaml.example` — N/A, no config keys added or changed
- [x] `CONTRIBUTING.md` / `AGENTS.md` — N/A, no architecture or workflow change
- [x] Cross-platform impact considered — the changes are prompt strings and env-var reads, no platform-specific code paths
- [x] I've updated tool descriptions/schemas — that's the substance of the `tools/terminal_tool.py` half

## Notes for review

- The `pinggy-tunnel` skill still uses `UserKnownHostsFile=/dev/null`, which is legitimate for ephemeral tunnel endpoints, so the new wording says "for a real host" rather than forbidding the flag outright. Rewriting that skill felt like a separate PR — happy to do it if you'd rather they read consistently.
- The description line ships in every request, so it costs tokens per turn. I trimmed it once already; if you'd still rather it were shorter, or split so only the `pty` half lands, say the word and I'll cut it.
- Not attempted here: letting an agent reach a *second* host mid-session. A `host` parameter on `terminal` routing through the existing `SSHEnvironment` (ControlMaster and file-sync are already there) would do it, but it widens the approval surface considerably and should be its own discussion.

https://claude.ai/code/session_01EB6DyhPVVoTgF1ZSAcnp9F

### DIFF
--- a/agent/prompt_builder.py
+++ b/agent/prompt_builder.py
@@ -1269,6 +1269,27 @@ def _probe_remote_backend(env_type: str) -> str | None:
     return formatted
 
 
+def _ssh_target_summary() -> str:
+    """Return ``user@host[:port]`` for the configured SSH backend, or ``""``.
+
+    Read from the same ``TERMINAL_SSH_*`` env vars ``_get_env_config`` uses
+    (config.yaml ``terminal.ssh_*`` keys are exported to these at startup),
+    so the hint names the machine the backend actually connects to. The probe
+    reports the remote OS/user/cwd but never *which* host — without this the
+    agent is told it is on some remote Linux box with no way to name it.
+    Destination only: the key path never enters the prompt.
+    """
+    host = (os.getenv("TERMINAL_SSH_HOST") or "").strip()
+    if not host:
+        return ""
+    user = (os.getenv("TERMINAL_SSH_USER") or "").strip()
+    target = f"{user}@{host}" if user else host
+    port = (os.getenv("TERMINAL_SSH_PORT") or "").strip()
+    if port and port != "22":
+        target = f"{target}:{port}"
+    return target
+
+
 def _clear_backend_probe_cache() -> None:
     """Test helper — drop the backend probe cache so monkeypatched backends take effect."""
     _BACKEND_PROBE_CACHE.clear()
@@ -1286,7 +1307,10 @@ def build_environment_hints() -> str:
       modal, daytona, ssh, vercel_sandbox): host info is **suppressed**
       because the agent's tools can't touch the host — only the backend
       matters. A live probe inside the backend reports its OS, user, $HOME,
-      and cwd. Falls back to a static summary if the probe fails.
+      and cwd. Falls back to a static summary if the probe fails. For the
+      **ssh** backend the configured ``user@host[:port]`` is named alongside
+      the backend, since the probe reports the remote OS/user/cwd but not
+      which host they belong to.
 
     The WSL environment hint is appended unchanged when running under WSL.
     """
@@ -1333,9 +1357,11 @@ def build_environment_hints() -> str:
     else:
         # --- Remote backend block (host info suppressed) ---
         probe = _probe_remote_backend(backend)
+        ssh_target = _ssh_target_summary() if backend == "ssh" else ""
+        backend_label = f"{backend} ({ssh_target})" if ssh_target else backend
         if probe:
             hints.append(
-                f"Terminal backend: {backend}. Your `terminal`, `read_file`, "
+                f"Terminal backend: {backend_label}. Your `terminal`, `read_file`, "
                 f"`write_file`, `patch`, and `search_files` tools all operate "
                 f"inside this {backend} environment — NOT on the machine "
                 f"where Hermes itself is running. The host OS, home, and cwd "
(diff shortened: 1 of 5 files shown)


Sheardo:
Closing — opened against the wrong repo. I was exploring hermes-agent for reference, not contributing to it. Apologies for the noise.

PR #98500, opened 2026-08-30
fix(state): audit-log end_session/reopen_session with the calling module

### LINKED ISSUE #98495: [Bug]: live gateway session row is transiently marked ended in state.db; #54878 stale-route self-heal fires repeatedly and silently reopens it
## Bug Description

The routing-time self-heal added for #54878 ("stale routing self-heal") fired **4 times in one day, across 2 gateway profiles — every single time against a session that was provably still live**. In each occurrence, state.db had a non-NULL `end_reason` on a session row that was actively serving turns, and the recovery path silently reopened it. No permanent end marker remains.

Signature (gateway.log):

```
WARNING gateway.session: gateway.session: routing key 'agent:main:telegram:dm:***' -> 20260830_175705_**** is ended in
state.db but still live in sessions.json; dropping stale entry and recovering/recreating the session (#54878)
```

This is the live-gateway variant of #54878, but the direction is inverted from what the original issue assumed: the session in state.db was **wrongly ended**, not the routing entry that was stale. `_is_session_ended_in_db()` (gateway/session.py:2479) reads `end_reason` as the liveness oracle, so any spurious `end_reason` write on a live row turns the next inbound message into a self-heal cycle.

Occurrences (2026-08-30, local time):

| Profile / platform | Session | Inbound that detected it | Row state after (queried ~20:00) |
|---|---|---|---|
| A / telegram DM | S1 | 18:21:43 | `end_reason = NULL`, 266 msgs |
| A / telegram DM | S1 (again) | 19:25:05 | `end_reason = NULL`, 266 msgs |
| B / google_chat DM | S2 | 10:22:00 | `end_reason = NULL`, 103 msgs |
| B / google_chat DM | S2 (again) | 17:18:30 | `end_reason = NULL`, 103 msgs |

S1 was marked ended **twice** within ~64 minutes. This rules out a one-off manual action; the writer is automated and repeatable.

### 1. The flagged sessions were live

S1 (telegram DM, profile A) was created at 17:57:05 by a compression split from its parent (parent row correctly ended with `end_reason='compression'`). API calls tagged `[S1]` ran continuously:

- 18:14:23 – 18:15:19 — turn in progress; 18:15:19 turn ended `interrupted_during_api_call` (user sent a new message mid-turn). The routing check on this inbound found S1 **alive**.
- 18:18:01 — next turn completed normally on S1 (`text_response`, 2 API calls, delivered 18:18:02).
- 18:21:43 — next inbound: self-heal fires, S1 found "ended in state.db".
- 18:21:4x — recovery reopens S1 (`reopen_session`, gateway/session.py:2894); the conversation continues on the **same** session_id.
- 19:31:45 – 19:36:29 — more turns on S1 after the second self-heal.

So the spurious `end_reason` write on S1 landed in the window **(18:15:19, 18:21:43)** — after the last routing check that observed S1 alive, before the next one. `end_session()` (hermes_state.py:7194) and `reopen_session()` (hermes_state.py:7212) are **not logged**, so the writer is not directly visible in current logs — only detectable through the next routing check.

### 2. Profile A's second self-heal triggered a cascading (but aborted) compression

At 19:25:05 (same inbound that hit the self-heal), session hygiene auto-compression also fired on an estimated ~274,755 tokens (the actual last API input for that session was ~131,800 tokens). The hygiene path's 10s turn-hold budget (`hygiene_max_turn_hold_seconds`, deliberately capped below the Telegram idle-sever window) was exhausted, the turn proceeded uncompressed, the detached compression pass then ran ~139.7 s and was commit-fenced:

```
INFO  gateway.run: Session hygiene compression for session ... exceeded the turn-hold budget (10.0s >= 10.0s)
      — abandoning inline wait, proceeding without compression this turn
WARNING gateway.run: Session hygiene auto-compress failed: turn-hold budget 10.0s elapsed after 10.0s
INFO  agent.conversation_compression: Compression commit cancelled before session mutation (session=...)
INFO  agent.conversation_compression: ... "failure_class":"commit_fence_cancelled" ... "total_duration_ms":139657
INFO  agent.turn_context: Skipping preflight compression: same-session cooldown active (~58 seconds remaining)
```

Net effect: a full long local-model compression pass was computed and discarded, and the log shows a cluster of scary-looking compression failures that are actually downstream of the stale-row false positive. (No user-visible outage: the turn completed and delivered normally.)

### 3. Same signature, second profile

Profile B (google_chat DM): same warning at 10:22:00 and 17:18:30 on session S2; S2's row is now `end_reason = NULL` with 103 messages and an intact compression parent chain.

## Prime suspect writer class (unconfirmed — no audit log exists)

The only runtime writer of a recoverable `end_reason` (`agent_close` / `ws_orphan_reap` / ... see `_RECOVERABLE_END_REASONS`, hermes_state_common.py:212) on a gateway session row is:

```
AIAgent.close() → session_db.end_session(session_id, "agent_close")   # run_agent.py:4752
```

gated by `_end_session_on_close`, which **defaults to True** (agent/agent_init.py:1770). Known fork paths that deliberately share a parent's *live* session_id opt out **individually, per call site**:

- `background_review.py:1304` — `review_agent._end_session_on_close = False`, with the comment that without the opt-out, close() "would finalize the parent's still-active session row mid conversation"
- `gateway/run.py` (hygiene compression agent), `gateway/slash_commands.py` (manual compression temp agent) — same opt-out

So the protection is a per-call-site flag, not structural: any `close()` on an AIAgent whose `session_id` is still the target of a live gateway routing entry will finalize that live row — and the only thing hiding the bug is the recovery path that quietly reopens it. Close sites that run **off the main turn path** and are therefore plausible writers: the cache-evict daemon threads spawned at `gateway/run.py:5863` (`agent-xproc-evict-*` / `agent-evict-*` — in profile A one such thread retired a cached agent's clients at 18:12:46, i.e. inside the suspect window for the second turn), and `request_error_cleanup` client teardown. Note the bg-review fork for S1 closed at 19:00:06 (correctly opted out), and the delegation subagent closes at 18:37:55 targeted their own session rows (verified: those subagent rows ended with their own `agent_close`, exactly 600 s after creation) — so those two are exonerated for this trace, leaving the evict/close paths unaccounted for.

## Proposed mitigations

1. **Audit-log `end_session()` and `reopen_session()`** — one line each: session_id, reason, and the caller (explicit `from=` parameter or stacklevel-2 module). Today these writes are silent, which is precisely why this bug is undiscoverable from logs. (cf. the "this was a silent logger.debug for months" comment on the failed-end warning at gateway/session.py:2995, #82616.)
2. **Find and fix the writer** that finalizes a live session row, or make `close()` refuse / loudly warn when the session_id is still the target of a live gateway routing entry.
3. **Error-amplify repeat self-heals**: count self-heal firings per (session_key, session_id); a second firing on the same live session within minutes is a bug, not a one-off stale route — escalate from WARNING to ERROR so it is visible in `errors.log`.

## Environment

- hermes-agent `89b38ed734` (v2026.8.27-533-g89b38ed734), Linux, two profiles each running `hermes-gateway` under systemd
- Main model and compression auxiliary: local vLLM endpoint (27B model, thinking disabled for compression, `reasoning_effort: none`)
- Compression: enabled, threshold 0.5; `context_total_ceiling_seconds: 300`
- Affected platforms: telegram DM (profile A), google_chat DM (profile B)

liuhao1024:
## What does this PR do?

Implements the audit-trail mitigation requested in #98495 (its "Proposed mitigations" item 1): `SessionDB.end_session()` and `SessionDB.reopen_session()` now emit one INFO line recording the session id, the end reason, and the **calling module.function** that performed the write.

Context: #98495 documents a live gateway session row being transiently marked ended (non-NULL `end_reason` on a row still serving turns) four times in one day across two profiles, silently reopened by the #54878 stale-route self-heal. The report itself states the writer is "unconfirmed — no audit log exists": both writes are completely invisible, so the bug can only be diagnosed from the downstream routing warning, never from the writer. I statically audited every gateway cleanup path that can reach `end_session()` (the three known shared-session callers already opt out via `_end_session_on_close = False`; evictions soft-release via `release_clients()`), and the remaining suspect class is a runtime race that cannot be pinpointed without exactly this trail. With this change, the next occurrence logs one line naming the culprit — e.g. `Session ended: session=... reason=agent_close caller=gateway.run._cleanup_agent_resources`.

Design notes:

- The caller is captured with `sys._getframe(1)` at method entry, before `_execute_write` adds its own frames between the caller and the SQL.
- The first-reason-wins no-op on an already-ended row logs at DEBUG, so the many legitimate double-close paths don't chatter at INFO.
- `reopen_session()` only logs the **ended → open transition** (the recovery resurrection); defensive reopens of already-live rows stay silent, and the row state read that decides this happens inside the same write transaction.

## Related Issue

Fixes #98495

## Type of Change

## Changes Made

- `hermes_state.py` — `end_session()`: INFO audit line (session/reason/caller) on the effective write, DEBUG no-op line when the row is already ended (first reason wins, unchanged semantics); `reopen_session()`: INFO audit line (session/caller) only when the row actually transitions from ended to open. SQL and write behavior unchanged — additive logging only.
- `tests/hermes_state/test_98495_end_reopen_audit_log.py` — new: pins the end audit line's session/reason/caller fields, the DEBUG level of the idempotent no-op (plus unchanged first-reason-wins row state), the reopen transition line, and the silence of reopens on live rows.

## How to Test

1. `python -m pytest tests/hermes_state/test_98495_end_reopen_audit_log.py -q` — should pass: 4 passed.
2. `python -m pytest tests/hermes_state/ tests/test_hermes_state.py tests/gateway/test_async_session_db.py tests/gateway/test_async_session_store.py tests/gateway/test_session_continuity_82616.py tests/gateway/test_session_store_stale_prune.py tests/gateway/test_dead_targets.py tests/cron/test_scheduler_cron_session_isolation.py tests/cron/test_scheduler.py tests/cron/test_scheduler_completion_verification.py -q` — Observed result: 594 passed, 2 skipped (all existing end/reopen consumers covered).
3. Manual smoke: in a Python shell, create a `SessionDB`, `create_session('s1', ...)` then `end_session('s1', 'agent_close')` with INFO logging enabled — observed line: `Session ended: session=s1 reason=agent_close caller=<module>.<function>`.

### Code

- [x] I've run `pytest tests/ -q` and all tests pass (full suites for `tests/hermes_state/`, `tests/test_hermes_state.py`, and the gateway/cron consumer subsets listed under How to Test: 594 passed, 2 skipped)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (method docstrings already describe the semantics; the new log lines are self-describing)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (pure logging, no OS-specific code)

## Screenshots / Logs

```
INFO hermes_state: Session ended: session=20260830_175705_abcd reason=agent_close caller=run_agent.close
INFO hermes_state: Session reopened: session=20260830_175705_abcd caller=gateway.session.get_or_create_session
```

### DIFF
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -7201,12 +7201,31 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
         with a different reason. Use ``reopen_session()`` first if you
         intentionally need to re-end a closed session with a new reason.
         """
+        # Audit trail (#98495): an end write on a still-live gateway session
+        # row is invisible today — the writer cannot be identified from logs,
+        # only inferred from the next routing self-heal warning. Record who
+        # ended what and why. Frame 1 is the direct caller; taken here because
+        # _execute_write runs _do synchronously but with extra frames in between.
+        _cf = sys._getframe(1)
+        _caller = f"{_cf.f_globals.get('__name__', '?')}.{_cf.f_code.co_name}"
+
         def _do(conn):
-            conn.execute(
+            updated = conn.execute(
                 "UPDATE sessions SET ended_at = ?, end_reason = ? "
                 "WHERE id = ? AND ended_at IS NULL",
                 (time.time(), end_reason, session_id),
             )
+            if updated.rowcount:
+                logger.info(
+                    "Session ended: session=%s reason=%s caller=%s",
+                    session_id, end_reason, _caller,
+                )
+            else:
+                logger.debug(
+                    "Session end no-op (already ended, first reason wins): "
+                    "session=%s reason=%s caller=%s",
+                    session_id, end_reason, _caller,
+                )
         self._execute_write(_do)
 
     def reopen_session(self, session_id: str) -> None:
@@ -7215,6 +7234,9 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
         Before clearing a reset boundary, stabilize markerless legacy reset
         children that still depend on the parent's mutable end_reason.
         """
+        _cf = sys._getframe(1)
+        _caller = f"{_cf.f_globals.get('__name__', '?')}.{_cf.f_code.co_name}"
+
         def _do(conn):
             placeholders = ",".join("?" for _ in _RESET_END_REASONS)
             # WHERE shape shared with _RESET_CHILD_SQL's fallback arm via
@@ -7230,10 +7252,22 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
                 f"AND {_legacy_reset_child_sql('child', placeholders)}",
                 (session_id, *_RESET_END_REASONS),
             )
+            _was_ended = conn.execute(
+                "SELECT ended_at FROM sessions WHERE id = ?",
+                (session_id,),
+            ).fetchone()
             conn.execute(
                 "UPDATE sessions SET ended_at = NULL, end_reason = NULL WHERE id = ?",
                 (session_id,),
             )
+            if _was_ended and _was_ended[0] is not None:
+                # Only the ended -> open transition is interesting: that is
+                # the stale-route recovery path silently resurrecting a row
+                # (#54878 self-heal), the counterpart of the end audit above.
+                logger.info(
+                    "Session reopened: session=%s caller=%s",
+                    session_id, _caller,
+                )
         self._execute_write(_do)
 
     def promote_to_session_reset(
--- a/tests/hermes_state/test_98495_end_reopen_audit_log.py
+++ b/tests/hermes_state/test_98495_end_reopen_audit_log.py
@@ -1,0 +1,76 @@
+"""Session end/reopen writes must identify their caller (#98495).
+
+A live gateway session row was transiently marked ended (end_reason set on a
+row still serving turns) and silently reopened by the #54878 stale-route
+self-heal — four times in one day across two profiles. Because
+``end_session()`` / ``reopen_session()`` logged nothing, the writer could
+not be identified from logs at all; only the downstream routing warning was
+visible. These tests pin the audit trail: every end write records
+session/reason/caller, the first-reason-wins no-op stays at DEBUG, and only
+the ended -> open transition of a reopen is worth an INFO line.
+"""
+import logging
+
+import pytest
+
+from hermes_state import SessionDB
+
+
+@pytest.fixture
+def db(tmp_path):
+    return SessionDB(tmp_path / "state.db")
+
+
+def test_end_session_logs_reason_and_caller(db, caplog):
+    db.create_session("s1", source="desktop")
+    with caplog.at_level(logging.INFO, logger="hermes_state"):
+        db.end_session("s1", "agent_close")
+    matches = [r for r in caplog.records if "Session ended" in r.message]
+    assert matches, f"expected an end audit line, got {[r.message for r in caplog.records]}"
+    line = matches[0].getMessage()
+    assert "session=s1" in line
+    assert "reason=agent_close" in line
+    # The caller is what makes the trail diagnostic: it must name the module
+    # and function that performed the write, not just the row that changed.
+    assert "test_end_session_logs_reason_and_caller" in line
+    assert __name__ in line
+
+
+def test_already_ended_row_logs_noop_at_debug(db, caplog):
+    db.create_session("s1", source="desktop")
+    db.end_session("s1", "compression")
+    with caplog.at_level(logging.DEBUG, logger="hermes_state"):
+        db.end_session("s1", "agent_close")
+    matches = [
+        r for r in caplog.records
+        if "Session end no-op" in r.message and r.levelno == logging.DEBUG
+    ]
+    assert matches
+    line = matches[0].getMessage()
+    assert "session=s1" in line
+    assert "reason=agent_close" in line
+    # First reason wins: the row was not re-ended under the new reason.
+    assert db.get_session("s1")["end_reason"] == "compression"
+
+
+def test_reopen_session_logs_the_ended_to_open_transition(db, caplog):
+    db.create_session("s1", source="desktop")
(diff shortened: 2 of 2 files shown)


Enough1122:

### PR #98500 — fix(state): audit-log end_session/reopen_session with the calling module

**Summary:** `SessionDB.end_session()` and `reopen_session()` (`hermes_state.py:7201`) now emit INFO audit lines `Session ended: session=… reason=… caller=module.func` / `Session reopened: session=… caller=…` via `sys._getframe(1)` captured before `_execute_write` indirection. Already-ended no-op stays at `DEBUG` ("Session end no-op (already ended, first reason wins)"), and reopen only logs the `ended → open` transition (live → open is silent). Tests in `tests/hermes_state/test_98495_end_reopen_audit_log.py` pin all three branches including caller naming.

- `sys._getframe(1)` inside `end_session`/`reopen_session` correctly names the direct caller (`test_*.py::test_*` in tests), not `_do`/`_execute_write` frames. If a helper wraps these, the helper's name would appear — acceptable for audit; consider `inspect.stack()[1]` is heavier and not needed.
- `rowcount` check for `end_session` and `SELECT ended_at` pre-check for `reopen_session` ensure only state-changing writes log at INFO — matches mitigations requested in #98495 and avoids spamming on defensive reopen paths.
- `logger` import already present; `sys`/`time` dependencies verified at module top — no new imports needed.

**Verdict:** Correct provenance fix for the transient live-row ended leak (#98495). No issues.

liuhao1024:
Thanks for the review! Confirming the details: the `sys._getframe(1)` capture happens before the `_execute_write` indirection so the audit line names the real caller rather than the write helper, and the `rowcount` / `ended_at` pre-checks are what keep INFO logging pinned to actual state transitions (the already-ended no-op stays at DEBUG and live→open reopens stay silent).

PR #98501, opened 2026-08-30
fix(tui): give the terminal client the desktop's todo-revision parity

nftpoetrist:
## Summary

The revisioned-snapshot/merge treatment desktop's todo panel got across several recent commits (`cf692532cf` merge-not-replace, `393af4a310` revisioned snapshots + `todo.updated` event, `58523f284c` unversioned-patch handling) never reached the `ui-tui` terminal client. Concretely, at current main:

- `createGatewayEventHandler.ts` has no `case 'todo.updated':` — the backend's dedicated full-snapshot event (`tui_gateway/server.py`'s `_on_tool_complete`, emitted specifically so "every client can reconcile immediately without interpreting provider text or partial merge arguments") is silently dropped.
- `turnController.recordTodos()` applies any update it receives with no revision watermark, so an out-of-order/stale update could clobber a newer one.
- `session.resume`/`session.activate` responses carry a `todo_state` snapshot (`tui_gateway/methods_session.py` via `_attach_todo_state`), but `useSessionLifecycle.ts` never reads it — resuming or switching to a session silently drops its live task list until the next tool event.

## Changes

Ports the same design `apps/desktop/src/store/todos.ts` already uses:

- `turnStore.ts`: adds a `todoRevision` watermark to `TurnState`.
- `turnController.ts`: `recordTodos()` gains an optional revision param and rejects a stale (lower) revision instead of applying it over newer state; a revision-less update (the `tool.start`-shaped patch case) always applies without moving the watermark, matching desktop's `acceptRevision()`. New `applyTodoSnapshot()` backs both the live `todo.updated` event and session restore — on an idle session it drops (instead of re-pinning) a still-active list, since that means the prior turn ended without a final update and the list is stale.
- `createGatewayEventHandler.ts`: adds the `todo.updated` case, and threads `payload.revision` through `tool.complete`.
- `useSessionLifecycle.ts`: `activateLiveSession`/`resumeById` now apply `r.todo_state` via `applyTodoSnapshot(..., running)`.
- `gatewayTypes.ts`: adds `revision` to the `tool.complete` payload, a `todo.updated` event variant, and `todo_state` on `SessionResumeResponse`/`SessionActivateResponse`.

## Test plan

- [x] New unit tests for the revision watermark and `applyTodoSnapshot()` (unrevisioned patch applies, increasing revisions apply, a stale revision is rejected, a live active snapshot applies, an idle-resume active snapshot is dropped, an unused-store snapshot is ignored) in `turnControllerTodos.test.ts`.
- [x] New `createGatewayEventHandler.test.ts` cases: `todo.updated` applies a snapshot with no prior `tool.start`; a stale `tool.complete` revision arriving after a newer `todo.updated` is rejected.
- [x] Mutation-verify: reverted the production changes and confirmed all 8 new tests fail with the expected assertions, then restored them and confirmed all pass.
- [x] `npx tsc --noEmit` clean.
- [x] Full `ui-tui` vitest suite: 1733 passed, 3 skipped, 1 pre-existing unrelated failure (`execFileNoThrow.test.ts`, a timing-sensitive subprocess test that fails independently of this change on this machine's Python).

## Competitor check

Searched `gh pr list`/`gh issue list` for "ui-tui todo", "todo.updated ui-tui", "turnController todo revision", "todo revision watermark" (state=all) — no open or duplicate PR touches `ui-tui`'s todo handling. The only related recent work is `#98326` (merged as commit `58523f284c`, already on `main` at this PR's base), which fixed an analogous unversioned-revision regression on the **desktop** side — this PR brings `ui-tui` up to the same design that fix reinforced, it doesn't touch the same files.

### DIFF
--- a/ui-tui/src/__tests__/createGatewayEventHandler.test.ts
+++ b/ui-tui/src/__tests__/createGatewayEventHandler.test.ts
@@ -183,6 +183,32 @@ describe('createGatewayEventHandler', () => {
     expect(getTurnState().todos).toEqual(todos)
   })
 
+  it('applies a todo.updated full-snapshot event even with no prior tool.start', () => {
+    const appended: Msg[] = []
+    const todos = [{ content: 'Boil water', id: 'boil', status: 'in_progress' }]
+    const onEvent = createGatewayEventHandler(buildCtx(appended))
+
+    onEvent({ payload: { revision: 1, todos }, type: 'todo.updated' } as any)
+
+    expect(getTurnState().todos).toEqual(todos)
+  })
+
+  it('rejects a stale tool.complete revision after a newer todo.updated already landed', () => {
+    const appended: Msg[] = []
+    const onEvent = createGatewayEventHandler(buildCtx(appended))
+
+    onEvent({
+      payload: { revision: 2, todos: [{ content: 'x', id: 'a', status: 'completed' }] },
+      type: 'todo.updated'
+    } as any)
+    onEvent({
+      payload: { name: 'todo', revision: 1, todos: [{ content: 'x', id: 'a', status: 'pending' }], tool_id: 't1' },
+      type: 'tool.complete'
+    } as any)
+
+    expect(getTurnState().todos).toEqual([{ content: 'x', id: 'a', status: 'completed' }])
+  })
+
   it('prints compaction progress status into the transcript', () => {
     const appended: Msg[] = []
     const ctx = buildCtx(appended)
--- a/ui-tui/src/__tests__/turnControllerTodos.test.ts
+++ b/ui-tui/src/__tests__/turnControllerTodos.test.ts
@@ -36,3 +36,69 @@ describe('turnController.recordTodos — preserves the parent field', () => {
     expect(getTurnState().todos).toEqual([{ content: 'x', id: 'a', status: 'pending' }])
   })
 })
+
+// A revision watermark (mirroring apps/desktop/src/store/todos.ts's
+// acceptRevision) must reject an out-of-order snapshot instead of letting a
+// late-arriving stale event undo a newer one already on screen.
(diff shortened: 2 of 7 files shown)


Enough1122:

### PR #98501 — fix(tui): give the terminal client the desktop's todo-revision parity

**Summary:** Brings desktop's revisioned todo snapshot/merge to `ui-tui`. Adds `todo.updated` case to `createGatewayEventHandler` (`ui-tui/src/app/createGatewayEventHandler.ts:132`), `tool.complete.revision` plumbing, `turnController` revision watermark `acceptTodoRevision()`/`recordTodos(value, revision)`/`applyTodoSnapshot(value, revision, running)` (`ui-tui/src/app/turnController.ts:166`), `todoRevision: null|number` in `turnStore`, and `todo_state` restore on `session.resume`/`session.activate` via `useSessionLifecycle`. Mirrors `apps/desktop/src/store/todos.ts` semantics: unrevisioned always applies, stale (< current) rejected, live snapshot or idle-completed list applies, idle still-active list dropped to avoid re-pinning stuck panel, empty+revision0 ignored as unused-store.

- `isTodoRevision` (`Number.isSafeInteger && >=0`) rejects floats/negatives — correct. `acceptTodoRevision` updates watermark only when `current !== revision` so duplicate revisions are idempotent.
- `applyTodoSnapshot` running=true for `todo.updated` (always live) and `running` param for session restore matches desktop's `restoreSessionTodosFromSnapshot` — dropping an idle `in_progress`/`pending` list prevents ghost panels after reconnect.
- Tests cover both `createGatewayEventHandler` full-snapshot without prior `tool.start` and stale-revision rejection, plus `turnController` `recordTodos`/`applyTodoSnapshot` edge cases — good parity coverage.

**Verdict:** Correct narrow parity fix; mirrors desktop behavior exactly. No blocking issues.

PR #86180, opened 2026-08-14
fix(model-switch): distinct-named custom providers stay reachable in the picker and keep their own extra_body

sfrmattos:
Closes #86179

## Summary

Two related bugs, found while tracking down why a second `providers:` entry pointing at the same endpoint (used to toggle vLLM's `chat_template_kwargs.enable_thinking` for a hybrid-thinking model) was practically unusable:

**1. Picker: distinctly-named `providers:` entries silently collapsed into one row**

`list_authenticated_providers()` groups same-endpoint `providers:` entries into one picker row so a provider with several models doesn't produce N near-duplicate rows (the "Palantir Claude 4.6/4.7 Opus" → "Palantir Claude" case, from #36998). Section 3 (`providers:`) grouped purely on connection identity (api_url, credential, api_mode, headers) — never on name — while section 4 (`custom_providers:`) already grouped on a version-stripped name prefix too. That asymmetry meant two `providers:` entries at the same endpoint with deliberately different names (e.g. "vLLM" / "vLLM No-Think") collapsed into a single row, and the second entry's `extra_body` became unreachable from the picker even though it resolved correctly via `--provider <name>` or a direct config edit.

Fix: promote the existing (cosmetic-only) version-suffix-stripping heuristic into section 3's grouping *key*, mirroring section 4. Two entries whose names share no separator/version pattern now stay distinct rows; entries whose names differ only by an embedded version number (the Palantir case) still fold into one row with multiple models. `extra_body`/`extra_headers` remain additional identity signals for the rarer case of identical names with different configs.

**2. Runtime: `extra_body` could apply to the wrong provider, or stop applying after a live switch**

Independent of the picker, `_merge_custom_provider_extra_body()` (agent/agent_init.py) matched a named custom provider's `extra_body` using `agent.provider`, which is always the bare canonical `"custom"` for every `providers:`/`custom_providers:` entry — it carries no per-entry identity. With two entries sharing (base_url, model), the matcher fell through to "first entry with a non-empty extra_body" and could silently apply the WRONG entry's extra_body regardless of which one was actually selected.

[...]

PR #98504, opened 2026-08-30
fix(model-switch): pick /model's extra_body by target model, not just name

nftpoetrist:
## Summary

`hermes_cli/model_switch.py`'s `switch_model()` — the pure function behind every `/model` surface (gateway, CLI, Telegram) — resolves the switched-to custom provider's `request_overrides` via:

```python
_cp_for_ro = _get_named_custom_provider(target_provider)
if _cp_for_ro:
    request_overrides = _custom_provider_request_overrides(_cp_for_ro) or None
```

`_get_named_custom_provider()` matches a `providers:`/`custom_providers:` entry by **name alone** and returns the **first** match. Two entries can legitimately share a name/`provider_key` while targeting different models with different `extra_body` — e.g. a local vLLM endpoint with two `custom_providers:` rows under the same name, one per model, each toggling `chat_template_kwargs.enable_thinking` differently. A `/model` switch to the second entry's model silently carried the **first** entry's `extra_body`.

This is the same bug class `agent_init._merge_custom_provider_extra_body` and `agent_runtime_helpers._apply_switched_provider_request_overrides` (added by `b10b27e6f9`) already guard against, via the shared `_custom_provider_extra_body_for_agent` matcher (provider identity + `base_url` + **model**) — its own docstring: *"Matching by name alone would let a different model selected at the same named endpoint inherit an extra_body configured for another model."* `hermes_cli/model_switch.py`'s own resolution never got the same treatment.

## Fix

`_get_named_custom_provider` is still used to resolve the entry's identity (it accepts bare aliases the strict matcher doesn't). The actual `extra_body` is now looked up via `_custom_provider_extra_body_for_agent`, scoped to that identity, the already-resolved `base_url`, and the model being switched to — reusing the established matcher instead of inventing a new one.

## Test plan

- [x] New regression test `test_switch_model_extra_body_matches_target_model_not_first_same_name_entry`: two `custom_providers` entries share a name/base_url with different `model`/`extra_body`; switching to the second entry's model must return that entry's `extra_body`, not the first's.
- [x] Mutation-verify: reverted `hermes_cli/model_switch.py` only, confirmed the new test fails (returns the first entry's `extra_body` instead of the target model's), restored and confirmed it passes.
- [x] `tests/hermes_cli/test_model_switch_custom_providers.py`, `tests/gateway/test_model_command_request_overrides.py`, `tests/hermes_cli/test_custom_provider_model_switch.py`, `tests/hermes_cli/test_user_providers_model_switch.py`, `tests/hermes_cli/test_model_switch_configured_provider_routing.py`: 99 passed.
- [x] `tests/hermes_cli/ -k "model_switch or custom_provider"`: 206 passed, 1 pre-existing unrelated skip.
- [x] `tests/gateway/ -k request_overrides` (excluding modules that fail to collect on this machine for an unrelated missing `aiohttp` dependency): 13 passed, 3 skipped.

## Competitor check

- **#98277** (merged today, ~5 hours before this branch) added `request_overrides` handling to this exact function as part of a larger salvage, but kept the name-only `_get_named_custom_provider` lookup — confirmed by reading current `main` directly and by mutation-verifying against it: the bug is present on `main` as of this PR's base commit, after `#98277` landed.
- **#86180** (open, last updated 2026-08-16 — predates `b10b27e6f9`, which added `agent_runtime_helpers._apply_switched_provider_request_overrides`) touches a different layer entirely: `list_authenticated_providers()`'s picker row-grouping (so two same-endpoint entries stay selectable as distinct rows) and `agent_init`'s build-time matcher. Its `hermes_cli/model_switch.py` hunk is exclusively the picker-grouping helpers (`_extra_body_identity`, `_group_display_prefix`) — verified via `gh pr diff` — it does not touch the `switch_model()` `request_overrides` block this PR changes.

### DIFF
--- a/hermes_cli/model_switch.py
+++ b/hermes_cli/model_switch.py
@@ -2189,15 +2189,33 @@ def switch_model(
     # ``extra_body`` such as chat_template_kwargs) so a ``/model`` switch to a
     # custom provider applies it on the gateway, matching the default-provider
     # path. resolve_runtime_provider surfaces these for named custom providers.
+    #
+    # Matched by provider identity, base_url, AND model — the same rule
+    # agent_init._merge_custom_provider_extra_body and
+    # agent_runtime_helpers._apply_switched_provider_request_overrides apply
+    # via the shared _custom_provider_extra_body_for_agent matcher. Two
+    # custom_providers entries can share a name/provider_key while targeting
+    # different models with different extra_body (e.g. a local vLLM endpoint
+    # toggling chat_template_kwargs per model); _get_named_custom_provider
+    # alone always returns the FIRST name match regardless of which model
+    # this switch actually targets, so it can't be used to pick extra_body.
     request_overrides = None
     try:
-        from hermes_cli.runtime_provider import (
-            _get_named_custom_provider,
-            _custom_provider_request_overrides,
-        )
+        from hermes_cli.runtime_provider import _get_named_custom_provider
+        from hermes_cli.config import load_config, get_compatible_custom_providers
+        from agent.agent_init import _custom_provider_extra_body_for_agent
+
         _cp_for_ro = _get_named_custom_provider(target_provider)
         if _cp_for_ro:
-            request_overrides = _custom_provider_request_overrides(_cp_for_ro) or None
+            _cp_identity = str(_cp_for_ro.get("provider_key") or _cp_for_ro.get("name") or "").strip()
+            _extra_body = _custom_provider_extra_body_for_agent(
+                provider=f"custom:{_cp_identity}",
+                model=new_model,
+                base_url=base_url,
+                custom_providers=get_compatible_custom_providers(load_config()),
+            )
+            if _extra_body:
+                request_overrides = {"extra_body": dict(_extra_body)}
     except Exception:
         request_overrides = None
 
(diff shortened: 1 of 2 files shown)


Enough1122:

### PR #98504 — fix(model-switch): pick /model's extra_body by target model, not just name

**Summary:** `hermes_cli/model_switch.py:2189` `switch_model()` previously resolved `request_overrides` via `_get_named_custom_provider(target_provider)` name-only lookup, always returning the first list entry regardless of target model. Now resolves via `get_compatible_custom_providers(load_config())` + shared matcher `_custom_provider_extra_body_for_agent(provider=f"custom:{identity}", model=new_model, base_url=base_url, ...)` matching on provider identity + `base_url` + `model` — same rule as `agent_init._merge_custom_provider_extra_body` and `agent_runtime_helpers._apply_switched_provider_request_overrides`. Covers the two-entries-share-name-and-base_url-but-different-`extra_body` case (e.g., local vLLM toggling `chat_template_kwargs.enable_thinking` per model). Test confirms switching to `model-b` carries `enable_thinking:false` not first entry's `true`.

- `_cp_for_ro` gate still via `_get_named_custom_provider` but identity/base_url/model discrimination now done by the `extra_body` matcher — note that if `custom_providers` has two entries differing only in `extra_body` but sharing base_url+model, first match wins; expected.
- `try/except: request_overrides=None` preserves fail-open on import/config errors — `/model` still switches model even if overrides cannot be resolved.
- Handler late-binds `load_config`/`get_compatible_custom_providers` at call time so the mock's `hermes_cli.config.load_config` delegate matches `runtime_provider`'s own delegate — test correctly notes both must see same entries.

**Verdict:** Correct provider+model-scoped override fix. No issues.

PR #98506, opened 2026-08-30
fix(memory): thread gateway user identity into OpenViking turn payloads

### LINKED ISSUE #98498: OpenViking memory provider drops the gateway user identity that Hermes threads into initialize()
### What happens

The OpenViking memory provider is the only memory plugin that discards the gateway user identity Hermes threads into `initialize()`. On a multi-user instance, preferences extracted from an unnamed message are all attributed to a single configured user, and the attribution is permanent.

### Where

`agent/agent_init.py` explicitly threads the identity, with a comment saying why:

```python
# Thread gateway user identity for per-user memory scoping
if agent._user_id:
    _init_kwargs["user_id"] = agent._user_id
if agent._user_name:
    _init_kwargs["user_name"] = agent._user_name
```

The other three memory providers consume it:

```
plugins/memory/hindsight/__init__.py:1628   self._user_id  = str(kwargs.get("user_id") or "").strip()
plugins/memory/honcho/__init__.py:513       runtime_user_peer_name=kwargs.get("user_id") or None
plugins/memory/mem0/__init__.py:356         self._user_id = configured or kwargs.get("user_id") or _DEFAULT_USER_ID
```

`plugins/memory/openviking/__init__.py` does not. In `initialize()` it takes the user from config only:

```python
self._user = settings["user"]   # resolved from config / OPENVIKING_USER
```

`kwargs["user_id"]`, `kwargs["user_id_alt"]` and `kwargs["user_name"]` are never read anywhere in the file.

### Why it matters

OpenViking's `preferences` schema writes to `{{ user }}/{{ topic }}.md`, and that `user` field is produced by the extraction model, not by the runtime. The model has nothing to go on: in the payload the plugin sends, only assistant messages carry an identity.

```python
def payload_message(role, parts):
    payload = {"role": role, "parts": parts}
    if role == "assistant" and assistant_peer_id:
        payload["peer_id"] = assistant_peer_id     # user messages carry nothing
    return payload
```

So the model infers the owner from the text. When someone states a preference without naming themselves — the common case — it falls back to the single configured OpenViking user. On a one-person instance this is invisible. On a shared instance (we run agencies with several agents on one Hermes), every unnamed preference lands under the same person. `preferences.user` has `merge_op: immutable`, so a wrong owner is permanent for that card.

### The server already supports the right mechanism

This is what makes it look like an oversight rather than a design choice. OpenViking resolves memory ownership from the **user-role** messages of the transcript:

```python
# openviking/session/memory/memory_isolation_handler.py
def _message_target_id(self, msg):
    if not self._is_peer_owner_message(msg):   # requires role == "user"
        return None
    peer_id = safe_peer_id(getattr(msg, "peer_id", None))
    ...

@staticmethod
def _is_peer_owner_message(msg):
    return getattr(msg, "role", None) == "user"
```

Because the plugin only ever sets `peer_id` on assistant messages, this path can never fire.

### Suggested fix

Attach the runtime speaker to user messages, the way the agent peer is already attached to assistant messages:

```python
self._user_id   = str(kwargs.get("user_id") or "").strip()
self._user_name = str(kwargs.get("user_name") or "").strip()
...
if role == "user" and self._user_id:
    payload["peer_id"] = safe_peer_id(self._user_id)
```

Keeping `X-OpenViking-User` as the tenant is important: it is the shared data space, and swapping it per speaker would silo `entities` per person, which is not wanted. Only the per-message owner needs to become runtime-derived.

### Environment

Hermes with `memory.provider: openviking`, OpenViking server v0.4.16, multi-user gateway instance.

### Workaround

Setting `peer_enabled: false` on the `preferences` schema keeps everything in one shared space, which avoids wrong-owner cards but gives up per-person scoping entirely.

liuhao1024:
## What does this PR do?

The OpenViking memory provider was the only memory plugin that discarded the gateway user identity Hermes threads into `initialize()` (`user_id`). Because the plugin only ever set `peer_id` on assistant messages, the server's ownership resolution — which keys off `role == "user"` messages — could never fire, and every preference extracted from an unnamed message was permanently attributed to the single configured tenant user on multi-user instances.

This PR threads the runtime identity through both payload construction paths:

- `initialize()` now reads `kwargs["user_id"]` (same pattern as the hindsight/mem0 providers) and stores it as `self._user_id`.
- `_messages_to_openviking_batch()` accepts a `user_peer_id` argument and attaches it to user-role messages, mirroring how `assistant_peer_id` is already attached to assistant messages.
- `_turn_batch_payload()` (the text fallback path used when the structured batch sync fails) does the same, so both paths behave identically.

The configured `X-OpenViking-User` tenant is deliberately untouched: it stays the shared data space (entities remain shared); only per-message ownership becomes runtime-derived, exactly as the issue suggests. When no gateway identity is present (single-user instances), payloads are byte-identical to before.

Note on `safe_peer_id`: the issue's suggested fix calls it, but that helper lives in the OpenViking server package, which this plugin does not import (it is a pure HTTP client). The plugin's existing convention for peer ids is `str(...).strip()` (see the `assistant_peer_id` handling), and the server re-normalizes incoming peer ids through its own `safe_peer_id` in `_message_target_id`, so stripping client-side is equivalent and dependency-free.

## Related Issue

Fixes #98498

## Type of Change

## Changes Made

- `plugins/memory/openviking/__init__.py`: initialize `_user_id` in `__init__`; read `kwargs["user_id"]` in `initialize()`; accept `user_peer_id` in `_messages_to_openviking_batch()` and attach it to user-role messages in the `payload_message` closure; attach `peer_id` to the user message in `_turn_batch_payload()`; pass `user_peer_id` from `sync_turn()`.
- `tests/plugins/memory/test_openviking_provider.py`: 5 new tests — initialize threading, both payload paths with identity, and both paths without identity (backward-compat anchor).

## How to Test

1. Run `pytest tests/plugins/memory/test_openviking_provider.py -q` — Observed result: 68 passed.
2. Run `pytest tests/plugins/memory/ tests/openviking_plugin/ -q` — Observed result: 399 passed, 3 skipped; the only failures are 6 pre-existing `test_hindsight_provider.py` failures that reproduce identically on unmodified upstream/main (verified by stashing this change and re-running).
3. Negative anchor: stashing only the plugin change makes the 3 new positive tests fail while the 2 backward-compat tests keep passing, confirming the tests pin the fix.

### Code

- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(memory):`)
- [x] I've run `pytest tests/ -q` and all tests pass (full openviking + memory-plugin suites green; the 6 hindsight failures are pre-existing on main and unrelated to this change)
- [x] I've tested on my platform: macOS 15.2 (arm64)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (inline comments document the tenant vs. runtime-identity distinction)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (pure payload construction, no OS-specific behavior)

## For New Skills

N/A

### DIFF
--- a/plugins/memory/openviking/__init__.py
+++ b/plugins/memory/openviking/__init__.py
@@ -2276,6 +2276,12 @@ class OpenVikingMemoryProvider(MemoryProvider):
         self._account = ""
         self._user = ""
         self._agent = ""
+        # Runtime gateway user identity threaded through initialize()
+        # (#98498). Distinct from self._user, the configured OpenViking tenant:
+        # the tenant stays the shared data space (X-OpenViking-User header)
+        # while this id marks per-message ownership so the server can scope
+        # preferences to the actual speaker.
+        self._user_id = ""
         self._session_id = ""
         self._turn_count = 0
         # Server-asserted user space for explicit-uid URIs (#91995). Key the
@@ -2794,6 +2800,7 @@ class OpenVikingMemoryProvider(MemoryProvider):
         self._env_refresh_enabled = True
         self._session_id = session_id
         self._turn_count = 0
+        self._user_id = str(kwargs.get("user_id") or "").strip()
         hermes_home = str(kwargs.get("hermes_home") or "").strip()
         if not hermes_home:
             try:
@@ -3189,18 +3196,19 @@ class OpenVikingMemoryProvider(MemoryProvider):
         return {"type": "text", "text": content}
 
     def _turn_batch_payload(self, user_content: str, assistant_content: str) -> Dict[str, Any]:
+        user_message: Dict[str, Any] = {
+            "role": "user",
+            "parts": [self._text_part(user_content)],
+        }
+        if self._user_id:
+            user_message["peer_id"] = self._user_id
         assistant_message: Dict[str, Any] = {
             "role": "assistant",
             "parts": [self._text_part(assistant_content)],
         }
         if self._agent:
             assistant_message["peer_id"] = self._agent
-        return {
-            "messages": [
-                {"role": "user", "parts": [self._text_part(user_content)]},
-                assistant_message,
-            ]
-        }
+        return {"messages": [user_message, assistant_message]}
 
     def _post_session_turn(
         self,
@@ -4489,9 +4497,11 @@ class OpenVikingMemoryProvider(MemoryProvider):
         messages: List[Dict[str, Any]],
         *,
         assistant_peer_id: str = "",
+        user_peer_id: str = "",
     ) -> List[Dict[str, Any]]:
         """Convert Hermes canonical messages into OpenViking batch payloads."""
         assistant_peer_id = str(assistant_peer_id or "").strip()
+        user_peer_id = str(user_peer_id or "").strip()
         tool_calls_by_id: Dict[str, Dict[str, Any]] = {}
         completed_tool_ids: set[str] = set()
         skipped_tool_ids: set[str] = set()
@@ -4527,6 +4537,8 @@ class OpenVikingMemoryProvider(MemoryProvider):
             payload: Dict[str, Any] = {"role": role, "parts": parts}
             if role == "assistant" and assistant_peer_id:
                 payload["peer_id"] = assistant_peer_id
+            elif role == "user" and user_peer_id:
+                payload["peer_id"] = user_peer_id
             return payload
 
         def flush_tool_parts() -> None:
@@ -4632,6 +4644,7 @@ class OpenVikingMemoryProvider(MemoryProvider):
         batch_messages = self._messages_to_openviking_batch(
             turn_messages,
             assistant_peer_id=getattr(self, "_agent", _DEFAULT_AGENT),
+            user_peer_id=getattr(self, "_user_id", ""),
         )
 
         if _sync_trace_enabled():
--- a/tests/plugins/memory/test_openviking_provider.py
+++ b/tests/plugins/memory/test_openviking_provider.py
@@ -1921,3 +1921,81 @@ class TestOpenVikingEnvWriter:
         assert env.read_text(encoding="utf-8").splitlines() == [
             "A=1", "OPENAI_API_KEY=new", "B=2",
         ]
+
+
+def test_initialize_threads_gateway_user_identity(monkeypatch):
+    """The gateway threads user_id into initialize() so memory providers can
+    scope per-user data (#98498). OpenViking used to drop it, attributing
+    every unnamed preference to the single configured tenant user."""
+    _clear_openviking_env(monkeypatch)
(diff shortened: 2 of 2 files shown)


Enough1122:

### PR #98506 — fix(memory): thread gateway user identity into OpenViking turn payloads

**Summary:** OpenViking was the only memory plugin discarding `user_id` threaded into `initialize()` via the gateway. Adds `self._user_id` (`plugins/memory/openviking/__init__.py:2279`) stored from `kwargs.get("user_id")`, distinct from `self._user` (OpenViking tenant / `X-OpenViking-User` header which remains the shared data space), and threads it as `peer_id` on user messages in `_turn_batch_payload` and in batched `_messages_to_openviking_batch(..., user_peer_id=...)` → `payload_message`'s `elif role=="user" and user_peer_id` branch. `sync_turn` passes `getattr(self, "_user_id", "")`. Enables server ownership resolution which keys off `role=="user"` messages, otherwise every unnamed preference collapsed to the single tenant user on multi-user instances.

- Empty `_user_id` keeps user messages bare (`"peer_id" not in`) — single-user instances unchanged. Tenant `self._user` untouched for V2 key resolution per `X-OpenViking-User`.
- Four new tests cover `initialize` threading (including no-identity stays `""`), `_turn_batch_payload` with/without identity, and batched multi-role payload — good.
- `initialize` fallback `str(kwargs.get("user_id") or "").strip()` handles `None`/whitespace correctly.

**Verdict:** Correct least-privilege ownership scoping for #98498. No blocking issues.

liuhao1024:
Thanks for the review — glad the ownership scoping reads correctly. The observations match the intent: empty `_user_id` keeps single-user payloads byte-identical, and the tenant `self._user` stays untouched for V2 key resolution. No blocking items to address; the PR stands as-is.

ehz0ah:
Thanks @liuhao1024. Your sender-attribution fix is now included in the OpenViking-owned external Hermes plugin through merged [OpenViking #5293](https://github.com/volcengine/OpenViking/pull/5293), adapted via #105812. Your original commit authorship is preserved.

The adaptation uses current per-turn author metadata so captures and retries retain the correct sender, while the configured OpenViking user stays unchanged. Mock gateway events passed capture → commit → extraction → recall checks against a real local OpenViking service.

This covers the external plugin; the bundled Hermes provider is unchanged.

liuhao1024:
Thanks @ehz0ah for the heads-up and for preserving the original authorship — glad to see the sender-attribution fix land in the OpenViking-owned plugin through #5293, and the per-turn author metadata adaptation matches exactly what this PR intended for captures and retries. Since this PR targets the bundled Hermes provider (which stays unchanged by that adaptation), I'll keep it open for maintainers to review.

PR #98508, opened 2026-08-30
fix(web): don't register web_extract when backend is search-only (ddgs/searxng/brave-free)

### LINKED ISSUE #89912: [Bug]: web_search registered but removed from live agent.tools with DDGS and local Ollama
### Bug Description

When using Hermes Agent with a local Ollama model and DDGS as the web-search backend, Hermes detects and registers both `web_search` and `web_extract`. However, the live agent passes only `web_extract` to the model API.

Consequently, the model correctly reports:

> The `web_search` tool is not available in this environment. The only web-access tool I have is `web_extract`.

DDGS is search-only, so attempts to substitute `web_extract` fail with:

```text
DuckDuckGo (ddgs) is a search-only backend
```

## Environment

* Hardware: NVIDIA DGX Spark, ARM64
* Hermes version: `v0.20.4 (2026.8.18)`
* Hermes commit: `395c70d6`
* Hermes installation: `/home/[user]/.hermes/hermes-agent`
* OpenAI Python SDK: `2.24.0`
* Model provider: Local Ollama via OpenAI-compatible API
* Ollama endpoint: `http://127.0.0.1:11434/v1`
* Model tested: `gpt-oss-hermes:latest`
* Search backend: DDGS
* Plugin: `web-ddgs` version `1.0.0`
* Tool Search: disabled

## Relevant configuration

```yaml
tools:
  tool_search:
    enabled: false

web:
  backend: ddgs
  search_backend: ""
  extract_backend: ""
  extract_char_limit: 15000
  use_gateway: false
```

No enabled or disabled toolset restrictions are configured:

```text
enabled_toolsets: None
disabled_toolsets: None
```

Hermes is launched with:

```bash
hermes chat --toolsets "web"
```

## Steps to reproduce

1. Install Hermes Agent v0.20.4.

2. Configure a local Ollama model through its OpenAI-compatible `/v1` endpoint.

3. Install DDGS in the Hermes virtual environment:

   ```bash
   cd ~/.hermes/hermes-agent
   ./venv/bin/python -m pip install --upgrade ddgs
   ```

4. Enable the DDGS provider:

   ```bash
   hermes plugins enable web-ddgs
   ```

5. Configure DDGS:

   ```yaml
   web:
     backend: ddgs
   ```

6. Disable Tool Search:

   ```yaml
   tools:
     tool_search:
       enabled: false
   ```

7. Start Hermes:

   ```bash
   hermes chat --toolsets "web"
   ```

8. Start a fresh session and prompt:

   ```text
   Call web_search once with query "DevelopKaki Malaysia".
   Return the raw results. Do not use web_extract.
   ```

### Expected Behavior

Hermes should pass both registered web tools to Ollama:

```text
['web_extract', 'web_search']
```

The model should issue a `web_search` tool call, and Hermes should execute it through DDGS.

### Actual Behavior

Hermes passes only:

```text
['web_extract']
```

The model therefore says that `web_search` is unavailable.

Because DDGS does not support extraction, subsequent `web_extract` attempts fail.

### Affected Component

Other, Tools (terminal, file ops, web, code execution, etc.)

N/A (CLI only)

### Debug Report

```shell
Automatic debug report withheld because it contains sensitive local configuration and log data.

Sanitised diagnostics:

Hermes Agent: v0.20.4 (2026.8.18)
Install directory: /home/[user]/.hermes/hermes-agent
OpenAI SDK: 2.24.0
OS: Ubuntu 24.04.4 LTS
Hardware: NVIDIA DGX Spark ARM64
Provider: Local Ollama
Endpoint: http://127.0.0.1:11434/v1
Model: gpt-oss-hermes:latest
Web backend: ddgs
Tool Search: disabled
Enabled toolset: web

Registered/final tool definitions:
['web_extract', 'web_search']

Live tools passed to the API:
['web_extract']

Sensitive configuration and complete logs are not included for security reasons. Additional sanitised diagnostic details can be provided on request.
```

### Operating System

Ubuntu 24.04.4 LTS

### Python Version

3.11.15

### Hermes Version

Hermes Agent v0.20.4

```shell
The following did not resolve the problem:

* Starting a completely fresh Hermes session
* Running Hermes outside the project workspace
* Disabling Tool Search
* Restricting the session to only the web toolset
* Changing between compatible Ollama models
* Changing the order of web_search and web_extract
* Verifying that no toolsets are disabled
* Changing the web_search registration check to:
=lambda: True
* Disabling Hermes’s between-turn MCP refresh
```

The tool snapshot is assigned around:

```text
agent/agent_init.py
```

Using:

```python
agent.tools = get_tool_definitions(...)
```

`get_tool_definitions()` independently returns both tools, but the live agent eventually reaches `build_api_kwargs()` with only `web_extract`.

The affected path appears to be between tool-definition resolution and the live `agent.tools`/`tools_for_api` snapshot used by the local Chat Completions transport.

## Impact

DDGS users with local Ollama models cannot use Hermes’s native `web_search` capability even though:

* DDGS is installed and operational
* The provider is enabled
* The tool is registered
* Tool assembly retains it
* The model supports it
* Ollama’s native and OpenAI-compatible endpoints support it

The only visible web tool becomes `web_extract`, which cannot operate with DDGS because DDGS is search-only.

### Proposed Fix (optional)

Please check:

1. How the CLI’s selected toolsets are transferred into the live `AIAgent`.
2. Whether `agent.tools` is rebuilt or filtered after initial registration.
3. Whether backend/provider capability filtering removes `web_search` incorrectly for keyless providers.
4. Whether plugin/provider registration timing differs between standalone `get_tool_definitions()` calls and live agent initialisation.
5. Whether the local Chat Completions path receives a stale or capability-filtered tool snapshot.

I can provide additional logs or test a proposed patch if required.

salch-cred:
## Problem

When \web.backend\ is set to a search-only provider (\ddgs\, \searxng\, \rave-free\), both \web_search\ and \web_extract\ share the same \check_fn=check_web_api_key\. Since DDGS makes that check return \True\, **both tools are registered** — but every call to \web_extract\ immediately fails with:

> \DuckDuckGo (ddgs) is a search-only backend\

This leaves the model in a broken state where it concludes \web_search\ is not available and falls back to worse extraction approaches.

Reported in: #89912

## Fix

- Add \_SEARCH_ONLY_BACKENDS\ frozenset (\ddgs\, \searxng\, \rave-free\).
- Add \check_web_search_available\ — search-only backends are valid for search; returns \True\ whenever \check_web_api_key\ does.
- Add \check_web_extract_available\ — returns \False\ when the shared backend is search-only **and** no \web.extract_backend\ override is configured.
- Wire the new per-capability \check_fn\ into both registry entries.

When \web.extract_backend\ is explicitly set (e.g. \ddgs\ for search + \irecrawl\ for extract), \web_extract\ is still registered — the split-backend configuration is preserved.

## Tests

\	ests/tools/test_web_search_only_backend.py\ — 7 parametrized cases covering:
- search-only backends: search available, extract not available
- search-only backend + explicit extract override: extract available
- full backend (firecrawl): extract available
- \_SEARCH_ONLY_BACKENDS\ contents invariant

Fixes #89912

### DIFF
--- a/tests/tools/test_web_search_only_backend.py
+++ b/tests/tools/test_web_search_only_backend.py
@@ -1,0 +1,74 @@
+"""Tests for search-only backend check_fn split (issue #89912).
+
+When web.backend is a search-only provider (ddgs, searxng, brave-free)
+and no per-capability extract_backend override is configured:
+- web_search must be available (check_fn returns True).
+- web_extract must NOT be available (check_fn returns False) so the model
+  is never shown a tool that always fails with "X is a search-only backend".
+
+When web.extract_backend is set explicitly, web_extract is available even if
+the shared backend is search-only.
+"""
+
+import pytest
+from unittest.mock import patch
+
+
+def _make_web_config(backend: str, extract_backend: str = "") -> dict:
+    return {"backend": backend, "extract_backend": extract_backend}
+
+
+@pytest.mark.parametrize("backend", ["ddgs", "searxng", "brave-free"])
+def test_search_only_backend_search_available(backend):
+    """web_search check_fn returns True for search-only backends."""
+    from tools.web_tools import check_web_search_available
+    with (
+        patch("tools.web_tools._load_web_config", return_value=_make_web_config(backend)),
+        patch("tools.web_tools._is_backend_available", return_value=True),
+    ):
+        assert check_web_search_available() is True
+
+
+@pytest.mark.parametrize("backend", ["ddgs", "searxng", "brave-free"])
+def test_search_only_backend_extract_not_available(backend):
+    """web_extract check_fn returns False when only a search-only backend is configured."""
+    from tools.web_tools import check_web_extract_available
+    with (
+        patch("tools.web_tools._load_web_config", return_value=_make_web_config(backend)),
+        patch("tools.web_tools._is_backend_available", return_value=True),
+    ):
+        assert check_web_extract_available() is False
+
+
+@pytest.mark.parametrize("backend", ["ddgs", "searxng", "brave-free"])
+def test_search_only_backend_with_extract_override_extract_available(backend):
+    """web_extract is available when extract_backend override is set even with search-only shared backend."""
+    from tools.web_tools import check_web_extract_available
+    with (
+        patch("tools.web_tools._load_web_config",
+              return_value=_make_web_config(backend, extract_backend="firecrawl")),
+        patch("tools.web_tools._is_backend_available", return_value=True),
+    ):
+        assert check_web_extract_available() is True
+
+
+def test_full_backend_extract_available():
+    """web_extract is available when a full (non-search-only) backend is configured."""
+    from tools.web_tools import check_web_extract_available
+    with (
+        patch("tools.web_tools._load_web_config", return_value=_make_web_config("firecrawl")),
+        patch("tools.web_tools._is_backend_available", return_value=True),
+    ):
+        assert check_web_extract_available() is True
+
+
+def test_search_only_constant_contents():
+    """_SEARCH_ONLY_BACKENDS contains exactly the known search-only built-in backends."""
+    from tools.web_tools import _SEARCH_ONLY_BACKENDS
+    assert "ddgs" in _SEARCH_ONLY_BACKENDS
+    assert "searxng" in _SEARCH_ONLY_BACKENDS
+    assert "brave-free" in _SEARCH_ONLY_BACKENDS
+    # Full backends must not be in the set
+    assert "firecrawl" not in _SEARCH_ONLY_BACKENDS
+    assert "exa" not in _SEARCH_ONLY_BACKENDS
+    assert "tavily" not in _SEARCH_ONLY_BACKENDS
--- a/tools/web_tools.py
+++ b/tools/web_tools.py
@@ -1525,6 +1525,25 @@ def _provider_is_ready(provider) -> bool:
         return False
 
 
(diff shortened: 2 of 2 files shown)


Enough1122:

### PR #98508 — fix(web): don't register web_extract when backend is search-only (ddgs/searxng/brave-free)

**Summary:** `tools/web_tools.py:1525` introduces `_SEARCH_ONLY_BACKENDS = frozenset({"ddgs","searxng","brave-free"})`, helper `_configured_backend_is_search_only()` (shared `backend` is search-only AND no `extract_backend` override), and splits the shared `check_fn=check_web_api_key` into `check_web_search_available()` (always `check_web_api_key()`) and `check_web_extract_available()` (False when search-only without override else `check_web_api_key()`). Rewires `web_search` and `web_extract` registry entries accordingly. Post-fix the model never sees a `web_extract` that would always fail with "X is a search-only backend", which previously led it to hallucinate "web_search is not available" (#89912).

- Case-folded `lower().strip()` on both `extract_backend` and `backend` so config capitalization does not bypass the gate. Empty override correctly treated as no-override.
- `web_search` deliberately stays a strict superset — search-only backends remain searchable; only extraction is hidden. With `web.extract_backend` override set, both tools are available regardless of shared backend — correct.
- Tests cover all three search-only backends for both tools, override case, full backends, and the constant contents — good coverage.

**Verdict:** Correct `check_fn` split fixing the model-visible broken tool. No issues.

PR #93858, opened 2026-08-24
feat(cron): make execution-ledger retention cap configurable (#93616)

OttoPrua:
High-volume profiles can exhaust the 1,000-row terminal execution ledger quickly, removing useful audit history for slower jobs. Independent production validation reported that one `*/5` job occupied 979 of 1,000 retained rows, leaving only 4 AM-report and 3 PM-report executions.

## What does this PR do?

This PR makes the terminal execution-ledger cap configurable per profile while preserving the existing default and pruning behavior:

- Reads `cron.max_terminal_executions` from the active profile's `config.yaml`.
- Keeps the backward-compatible default of 1,000 terminal rows.
- Validates values as positive whole numbers within SQLite's signed 64-bit integer range.
- Fails closed when the cap is invalid or cap resolution raises: the terminal execution state is committed, but destructive pruning is skipped.
- Logs the first error for each distinct invalid setting and ledger, suppressing repeated high-volume error spam until a valid resolution resets the streak.
- Preserves the module-level test/caller override, including an explicit override equal to the default.
- Never prunes `claimed` or `running` executions.

This does not change cron cadence, scheduling, job execution, or per-job retention priority.

## Related Issue

Fixes #93616

Related implementations: #97818 uses a different configuration name and fallback policy; #98511 additionally introduces per-job fairness. This PR intentionally keeps the scope to the configurable, strictly validated global cap requested in #93616.

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [x] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `cron/executions.py`
  - Resolves and strictly validates the profile-local terminal-row cap before binding it to SQLite.
  - Skips pruning on invalid values without rolling back the terminal execution update.
  - Keeps override precedence, profile isolation, in-flight preservation, and deduplicated error reporting.
- `hermes_cli/config_defaults.py`
  - Registers `cron.max_terminal_executions` with a default of 1,000.
- `tests/cron/test_execution_ledger.py`
  - Covers valid, invalid, and oversized values; config and override precedence; profile isolation; fail-closed terminal writes; log deduplication/reset; and recovery pruning.
- `website/docs/user-guide/features/cron.md`
[...]

PR #98511, opened 2026-08-30
feat(cron): retain execution history fairly per job

### LINKED ISSUE #93616: Cron execution history pruned to 1000 rows (~36h at high job volume) — make MAX_TERMINAL_EXECUTIONS configurable
### Problem

`cron/executions.py` hardcodes the terminal-execution retention cap:

```python
MAX_TERMINAL_EXECUTIONS = 1000
```

and `_prune_unlocked()` deletes oldest rows beyond that cap on every write:

```sql
DELETE FROM executions WHERE id IN (
  SELECT id FROM executions
  WHERE status IN ('completed','failed','unknown')
  ORDER BY claimed_at DESC, id DESC LIMIT -1 OFFSET ?
)
```

There is no environment-variable or configuration hook for this value.

### Impact

Deployments running many short-interval jobs churn through 1000 rows very quickly. In our production instance (~670 executions/day across several `every 5m`/`*/10` monitors plus a `*/10` backfill job) the entire executions ledger covers only **~36 hours**, which makes cron-failure auditing and incident forensics impractical. Lower-volume instances keep ~3 weeks, so behavior is highly deployment-dependent.

Notably, upstream's own test (`tests/cron/test_execution_ledger.py`, `monkeypatch.setattr(executions, "MAX_TERMINAL_EXECUTIONS", 3)`) already treats the cap as an injection point — it just isn't reachable from configuration.

### Evidence at tag v2026.8.3

- https://github.com/NousResearch/hermes-agent/blob/v2026.8.3/cron/executions.py (constant at line 21, `_prune_unlocked` at line 123)
- https://github.com/NousResearch/hermes-agent/blob/v2026.8.3/tests/cron/test_execution_ledger.py

The prune logic is unchanged at v2026.8.19.

### Proposal

Make the budget configurable with backward-compatible default 1000, e.g.:

- env override: `HERMES_CRON_MAX_EXECUTIONS` (matches existing `HERMES_CRON_*` convention such as `HERMES_CRON_TIMEOUT`)
- and/or a config key under the cron scope (e.g. `cron.execution_retention_rows`) registered in `hermes_cli/config_defaults.py`, resolved at import/use time in `cron/executions.py`
- extend `tests/cron/test_execution_ledger.py` to cover the override paths

Happy to open a PR along these lines if maintainers agree on the mechanism.

tachyon-r:
## Summary

- make terminal execution retention configurable with `cron.execution_retention` while preserving the default hard cap of 1000 rows
- add `cron.execution_retention_per_job` as a fair-share target within that cap
- retain terminal rows by priority: latest attempt per represented job, failed attempts, fair-share rounds, then global recency
- keep claimed and running attempts outside terminal pruning
- preserve `MAX_TERMINAL_EXECUTIONS` as a test and embedding override
- document both settings and use explicit UTF-8 in configuration fixtures

Fixes #93616.

## Behavior

The cap remains strict even when represented jobs, failures, or fair-share targets exceed it. Protection tiers are priorities, not unbounded exemptions. A retention value of zero removes terminal rows while leaving active attempts untouched.

## Verification

- execution-ledger tests — 30 passed
- configuration tests — 112 passed, 4 platform-specific skips
- Ruff — passed
- Ty for `cron/executions.py` — passed
- Windows-footgun scan — passed
- `git diff --check` — passed

## Compatibility

No schema migration or new environment variable. Existing default capacity and the module-level override remain supported.

### DIFF
--- a/cron/executions.py
+++ b/cron/executions.py
@@ -24,7 +24,11 @@ from hermes_time import now as _hermes_now
 # that temporarily enter another profile cannot leak that profile's records into the import-time
 # home.
 EXECUTIONS_FILE: Optional[Path] = None
-MAX_TERMINAL_EXECUTIONS = 1000
+_DEFAULT_MAX_TERMINAL_EXECUTIONS = 1000
+# Backward-compatible test override. Production reads
+# ``cron.execution_retention`` at prune time so a config change does not
+# require a Python reinstall.
+MAX_TERMINAL_EXECUTIONS = _DEFAULT_MAX_TERMINAL_EXECUTIONS
 HANDOFF_ADOPTION_GRACE_SECONDS = 30.0
 _TERMINAL_STATES = ("completed", "failed", "unknown")
 _lock = threading.RLock()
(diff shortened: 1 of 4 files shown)


Enough1122:

### PR #98511 — feat(cron): retain execution history fairly per job

**Summary:** Makes terminal `cron/executions.db` pruning configurable and fair. Adds `cron.execution_retention` (default 1000, hard cap) and `cron.execution_retention_per_job` (default 5, fair-share target) in `hermes_cli/config_defaults.py`, helpers `_terminal_execution_retention()` / `_terminal_execution_retention_per_job()` reading profile-local `load_config_readonly()` (with `MAX_TERMINAL_EXECUTIONS` module override winning for tests), and a prioritized `WITH ranked AS (ROW_NUMBER PARTITION BY job_id) / keepers` SQL in `_prune_unlocked()`: latest-per-job (rank=1), then failed, then per-job-round-robin, then global recency, bounded by `LIMIT ?` hard cap. Claimed/running rows remain outside pruning. Docs and 8 new tests including hard-invariant parametrization `(0,1)-(11,20)`.

- `per_job` clamped to `max(1, int(value))` — 0 becomes 1, but `total` can be 0 (removes terminal rows after each write); `test_terminal_retention_cap_is_a_hard_invariant` confirms `terminal <= total` even when protected sets exceed cap, so tiers are priorities not exemptions — important correctness property.
- SQL ordering: `CASE WHEN job_rank<=per_job THEN job_rank ELSE NULL END` ensures per-job rounds are round-robin by rank, not recency-biased within the tier. `claimed_at DESC, id DESC` final tie-break is stable.
- Config defaults asserted equal module defaults in `test_config_default_matches_module_default` — prevents drift between docs and code.

**Verdict:** Well-documented, well-tested fairness improvement. No blocking issues.

tachyon-r:
Rebased onto pinned main bc1330eebc0aa8a443501b5f62586eb7361353a5. A behavioral test on that main confirmed that a frequent job still evicts a slow job under the terminal cap; the same test passes on this branch. Fair retention remains distinct from #93858. Focused validation through scripts/run_tests.sh: 157 passed, 4 native-platform skips. Fresh-head CI is green for 8f11de04ee7a39da7919996e672afcba3d06da32, including All required checks pass. Original fork branch was updated with an explicit original-head lease; no merge.

PR #91540, opened 2026-08-21
fix(desktop): pair Windows SSH runtime executables

1052326311:
## What does this PR do?

Fixes the Windows Desktop SSH probe when the official installer puts the launcher shim at `hermes-agent\\bin\\hermes.exe` ahead of the venv on `PATH`. The probe now validates `hermes.exe` and sibling `python.exe` as a pair before selecting a candidate, so it skips the shim and falls through to the installed venv runtime.

It also extracts PowerShell CLIXML error records before surfacing probe failures, so progress records no longer hide the actionable error.

Current base: `29033a3fd5830d1b0eb1aa271dead8d5e1580e8b`
Current head: `6261abcf1972fd3464f43c228a9a5d4da448a406`

The current probe preserves upstream no-reparse hardening: both executables must be non-reparse leaf files before a candidate pair is selected.

## Related Issue

Fixes #91479

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✅ Tests (adding or improving test coverage)

## Changes Made

- Validate each Windows Hermes candidate together with its sibling Python runtime before selection.
- Preserve explicit-path precedence while rejecting an explicit executable that is not paired with Python.
- Decode CLIXML `Error` records and PowerShell escaped newlines before rendering the unsupported-platform detail.
- Add regression coverage for paired-runtime selection and readable CLIXML failures.

## How to Test

1. On the pre-fix base, inspect the generated probe: it selects the first existing `hermes.exe`, then derives `python.exe`; the official `bin` shim therefore wins and points to a nonexistent sibling.
2. On this head, the generated probe filters candidates by both executable paths before selection and falls through to `venv\\Scripts`.
3. Current-head validation recorded in [the existing evidence comment](https://github.com/NousResearch/hermes-agent/pull/91540#issuecomment-5434685411): focused Electron suite **16/16**, scoped ESLint with zero warnings, Prettier check, Electron TypeScript check, diff check, and two-file TruffleHog scan (0 verified/unverified findings).
4. This body corrects the previously listed obsolete head and 8-test receipt; these are the existing August 27 results, not a new test run.

Windows hardware was not available locally; the regression validates the exact encoded PowerShell sent over SSH and the user-visible parsing boundary without opening a network connection.

### Code

- [x] Read the contributing and Desktop engineering guides
- [x] Commit follows Conventional Commits
[...]

Issue #91479, opened 2026-08-21
[Bug]: Desktop SSH: OS probe fails on Windows remotes — assumes python.exe next to hermes.exe


AkiraAsano-kobe-u:
### Bug Description

Adding an SSH connection to a **Windows** remote host and clicking **Test** always fails with:

The remote operating system is not supported by Desktop SSH.
(probe: SSH error connecting to <host>: #< CLIXML <Objs Version="1.1.0.1" xmlns="http://schemas.microsoft.com/powershell/2004/04">...)

Two separate problems are visible here:

1. **The message is misleading.** The `#< CLIXML` payload is PowerShell's *module auto-load progress* stream, not an error. It sends you looking in the wrong direction entirely. The real error is buried inside the XML in `<S S="Error">`:

   The remote Hermes Python runtime was not found.

2. **The probe cannot succeed on a stock Windows install.** `probeWindowsRemote()` assumes `python.exe` is a sibling of the resolved `hermes.exe`. That holds for `venv/Scripts/`, but the official installer also creates `hermes-agent/bin/` — containing only `hermes.exe` and `hermes-acp.exe`, no `python.exe` — and puts it **first on PATH**.

Because this is the installer's own layout, **every Windows → Windows Desktop SSH connection hits this.** It is not an unusual configuration.

### Steps to Reproduce

1. Install Hermes on a **Windows** host with the official installer:
   `iex (irm https://hermes-agent.nousresearch.com/install.ps1)`
2. Confirm the installer put `hermes-agent\bin` first on the user PATH, and that it contains only `hermes.exe` / `hermes-acp.exe` (no `python.exe`):
   ```powershell
   [Environment]::GetEnvironmentVariable("PATH","User") -split ';' | Where-Object { $_ -match 'hermes' }
   Get-ChildItem "$env:LOCALAPPDATA\hermes\hermes-agent\bin"
3. Enable OpenSSH Server on that host and set up key auth.
4. From another machine running Hermes Desktop: Settings → Connections → Add connection → SSH, enter the host, save.
5. Click Test.

Result: The remote operating system is not supported by Desktop SSH.

The underlying failure can be reproduced directly by running the probe script by hand:

$cmd = Get-Command hermes.exe -ErrorAction SilentlyContinue
$hermes = $cmd.Source                                  # -> ...\hermes-agent\bin\hermes.exe
$python = Join-Path (Split-Path $hermes) "python.exe"  # -> ...\hermes-agent\bin\python.exe
[...]

PR #98515, opened 2026-08-30
fix(desktop): avoid Windows SSH probe command-line limit

thegreatesthoneybee:
## Summary

- Stream the Windows Desktop SSH platform probe over SSH stdin instead of embedding a long PowerShell payload in the remote command line.
- Encode the streamed payload as UTF-16LE Base64 so Unicode paths survive the stdin transport.
- Keep probe statements CRLF-separated so PowerShell `try`/`catch` blocks parse correctly.
- Add transport, Unicode payload, and real-Windows-PowerShell regression coverage.

## Root cause

Windows OpenSSH commonly invokes `cmd.exe`, whose command-line limit rejects the UTF-16LE Base64 `-EncodedCommand` used by the probe. The actual error was `The command line is too long`, but Desktop surfaced it as `Unsupported remote platform`.

The probe now invokes a short PowerShell command and reads a Base64-encoded UTF-16LE script from stdin, keeping the payload out of the command line while preserving path characters.

## Related issues and PRs

- Related to #91479
- Complements #91540 (Windows runtime-pair and CLIXML handling)

## Test plan

- `windows-remote-lifecycle.test.ts`: 17/17 passed on Windows.
- Real PowerShell execution test runs through local `cmd.exe` and verifies the returned Windows probe JSON.
- Direct `cmd.exe`/PowerShell smoke test verifies a Unicode path payload returns `unicode-ok`.
- ESLint on both changed files: 0 errors, 0 warnings.
- `git diff --check`: passed.

The repository checkout is missing several UI/Electron dependencies, so the full repository `typecheck` is not currently runnable; unrelated baseline errors include missing Electron declarations and Windows-vs-POSIX assumptions in `ssh-connection.test.ts`.

## Scope

Only these files are changed:

- `apps/desktop/electron/windows-remote-lifecycle.ts`
- `apps/desktop/electron/windows-remote-lifecycle.test.ts`

### DIFF
--- a/apps/desktop/electron/windows-remote-lifecycle.test.ts
+++ b/apps/desktop/electron/windows-remote-lifecycle.test.ts
@@ -1,6 +1,6 @@
 import assert from 'node:assert/strict'
-import crypto from 'node:crypto'
 import { spawn } from 'node:child_process'
+import crypto from 'node:crypto'
 import fs from 'node:fs'
 import os from 'node:os'
 import path from 'node:path'
@@ -74,7 +74,11 @@ type SshExecOptions = { stdinData?: string | Buffer }
 function scriptFromInvocation(command: string, options: SshExecOptions = {}) {
   const encoded = command.match(/-EncodedCommand\s+([^\s]+)$/)?.[1]
 
-  return encoded ? Buffer.from(encoded, 'base64').toString('utf16le') : String(options.stdinData || '')
+  if (encoded) {
+    return Buffer.from(encoded, 'base64').toString('utf16le')
+  }
+
+  return Buffer.from(String(options.stdinData || '').trim(), 'base64').toString('utf16le')
 }
 
 test('PowerShell transport uses UTF-16LE encoded commands and literal escaping', () => {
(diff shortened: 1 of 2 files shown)


Enough1122:

### PR #98515 — fix(desktop): avoid Windows SSH probe command-line limit

**Summary:** Streams long Windows probe PowerShell over SSH `stdin` (`[ScriptBlock]::Create([Text.Encoding]::Unicode.GetString([Convert]::FromBase64String([Console]::In.ReadToEnd()))).Invoke()`) instead of embedding UTF-16LE Base64 in `cmd.exe` command line (8191-char limit). Script now CRLF-joined for correct `try`/`catch` parsing; payload encoded UTF-16LE Base64 preserves Unicode paths.

- New `powerShellStdinCommand()` is fixed string; `encodedPowerShell(script)` still does UTF-16LE to Base64 — correctly handles non-ASCII paths (covered by Unicode test).
- Tests add `scriptFromInvocation` helper supporting both `-EncodedCommand` legacy and `stdinData` new path; existing relaunch-gate tests updated to use it — no regression.
- `runLocalWindowsCommand` uses `spawn(cmd.exe)` and pipes `stdinData`; skipped on non-win32. Ensure `SshConnection.exec` indeed forwards `options.stdinData` to remote stdin (existing API already supported per comment).
- `command.length < 1024` assertion ensures probe stays well under limit.

**Verdict:** Correct fix for Windows OpenSSH cmd.exe limit. No blocking issues.

PR #83675, opened 2026-08-11
fix(tui): accelerate mechanical wheel scrolling in native terminals

fangliquanflq:
## What does this PR do?

Native terminal users with mechanical mouse wheels currently remain stuck at one row per notch, even during sustained fast scrolling through long TUI sessions. This change lets ordinary same-direction wheel events build bounded acceleration while preserving deliberate slow clicks, direction handling, trackpad behavior, and the separate xterm.js path.

### Symptom

On native terminals, repeatedly scrolling a mechanical wheel in one direction at a realistic cadence moves only one row per event. Faster wheel input therefore does not improve navigation speed through long transcripts.

### Impact

Users of native terminals with mechanical mouse wheels must issue many more wheel events to traverse long TUI output. The affected path is limited to native terminal wheel handling; xterm.js uses a separate acceleration path.

### Bug Cause

**Trigger:** `ui-tui/src/lib/wheelAccel.ts:149` / ordinary native same-direction wheel events whose gaps exceed 40 ms.

**Causal chain:**
1. A mechanical wheel emits sustained same-direction events at about 80 ms intervals.
2. The native wheel state machine treats every gap above 40 ms as idle and resets the multiplier to its base value.
3. Each event remains one row, so normal mechanical-wheel scrolling never engages acceleration unless encoder bounce happens to activate the alternate wheel mode.

**Why it is wrong:** The ordinary native acceleration window was narrower than realistic mechanical-wheel cadence, making its ramp unreachable for healthy same-direction wheel input.

**Working sibling / contrast:** The xterm.js path already uses its own cadence-aware decay model and is unchanged. Native encoder-bounce wheel mode can accelerate, but requiring a direction flip and flip-back is not a valid prerequisite for normal same-direction scrolling.

**Ruled out:** The native call site already forwards each decoded direction and timestamp into this pure state machine, so PTY transport and event decoding are not responsible for the multiplier reset.

### Fix

Widen the bounded ordinary-native acceleration window to 200 ms, matching the existing native mechanical-wheel cadence boundary. Add a regression test that models 80 ms same-direction events, verifies that they exceed one row without entering wheel mode, and keeps the six-row cap intact.

## Related Issue

Closes #83651

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `ui-tui/src/lib/wheelAccel.ts` - allow sustained native wheel events within 200 ms to build bounded acceleration.
[...]

PR #98517, opened 2026-08-30
fix(tui): add sustained-direction ramp to wheelAccel for quantised terminals

### LINKED ISSUE #98295: [Bug]: Can't use wheelAccel in modern terminal
### Bug Description

wheelAccel assumes that the terminal emulator sends touchpad events at a high frequency. However, in modern terminal emulators, the terminal sends events based on the distance of the touchpad. $120$ units in Windows 

### Steps to Reproduce

1. Open a long chat transcript in Hermes TUI: `hermes --tui` (on Windows Terminal, Ghostty, iTerm2, or any native terminal).
2. Spin a mechanical mouse wheel rapidly in one direction, or scroll smoothly on a precision touchpad.
3. Observe that scrolling speed never ramps up—it is locked at exactly 1 line per event regardless of speed.

### Expected Behavior

ustained scrolling in the same direction should smoothly ramp up acceleration (e.g., 2–6+ lines per tick) based on continuous momentum and decay over time.

### Actual Behavior

Every event scrolls exactly 1 line. Acceleration never engages.

### Affected Component

CLI (interactive chat)

_No response_

### Debug Report

```shell
None
```

### Operating System

Windows11

### Python Version

_No response_

### Hermes Version

_No response_

```shell

```

```
 In ui-tui/src/lib/wheelAccel.ts (nativeStep function):

 1. Unreachable wheelMode for normal hardware:
    ```ts
if (dir !== state.dir && state.dir !== 0) {
  state.pendingFlip = true
  state.time = now
  return 0
}
    ```
    wheelMode is only activated when a direction reversal flips back to the original direction within WHEEL_BOUNCE_GAP_MAX_MS (200ms). A functional mechanical mouse wheel rolling
    in one direction never emits reverse bounce events.
 2. Terminal Grid Quantization breaks timing assumptions:
    Native terminal emulators (e.g. Windows Terminal via _mouseWheelDelta, iTerm2 via _accumulatedScrollDeltaY, Alacritty via size_info.cell_height()) accumulate continuous
    pixel/touchpad deltas and only emit SGR mouse sequences when an entire character line is crossed. A steady touchpad swipe emits events spaced by ≈ 80ms ∼ 200ms.
 3. Hard 40ms cutoff and lack of fractional carry in nativeStep:
    ```ts
if (gap > WHEEL_ACCEL_WINDOW_MS) { // 40ms
  state.mult = state.base
} else {
  state.mult = Math.min(cap, state.mult + WHEEL_ACCEL_STEP) // +0.3
}

```
### Proposed Fix (optional)
```
 1. Unify around an Exponential Decay Model with Fractional Carry:
    Deprecate the brittle "Touchpad vs. Mechanical Wheel" heuristic in nativeStep and adopt a momentum-decay approach similar to xtermJsStep:
   - Use an exponential decay curve with a half-life of ≈ 150ms ∼ 250ms (WHEEL_DECAY_HALFLIFE_MS).
   - Allow consecutive same-direction events within 200ms ∼ 400ms to build acceleration momentum.
   - Introduce state.frac (fractional accumulator) into the native path so increments like +0.3 accumulate across frames (yielding 1, 2, 1, 2 rows) instead of being dropped by
     Math.floor.
 2. Remove Encoder Bounce Dependency:
    Direction reversals should simply reset the momentum multiplier to base (1), rather than gating acceleration behind a bounce detection window.

```

salch-cred:
## Problem

Modern terminals (Windows Terminal, Ghostty, iTerm2, Alacritty) accumulate touchpad/mouse-wheel deltas and only emit a SGR mouse event when a full character cell is crossed. Typical inter-event gaps are **80–300 ms** — well above the old \WHEEL_ACCEL_WINDOW_MS = 40 ms\ fast-burst threshold.

Before this fix every such event hard-reset the multiplier to base (1 row/event), so \wheelAccel\ acceleration **never engaged** during normal scrolling, regardless of scroll speed.

Reported in: #98295

## Fix

Add a **sustained-direction window** (40 ms – 500 ms) that applies a gentle half-life decay + grow step, keeping the multiplier elevated across quantised-delta events:

| Gap | Behaviour |
|-----|-----------|
| < 5 ms | same-batch burst → 1 row/event |
| 5–40 ms | fast-burst ramp (+0.5/event, cap 6) — unchanged |
| 40–500 ms | sustained ramp: decay × ½^(gap/200ms) + 0.15 grow |
| ≥ 500 ms | hard reset to base — deliberate single clicks stay crisp |

\WHEEL_ACCEL_STEP\ tuned 0.3 → 0.5 for the narrower fast-burst window.

## Tests

Two new cases in \src/__tests__/wheelAccel.test.ts\:
- *sustained scroll at ~120ms intervals ramps above base* — regression witness for #98295
- *sustained scroll stays within cap* — safety bound (≤ 6 rows/event)

All **15 wheelAccel tests pass**.

Fixes #98295

### DIFF
--- a/ui-tui/src/__tests__/wheelAccel.test.ts
+++ b/ui-tui/src/__tests__/wheelAccel.test.ts
@@ -20,14 +20,44 @@ describe('wheelAccel — native path', () => {
     expect(computeWheelStep(s, 1, 1060)).toBeGreaterThanOrEqual(1)
   })
 
-  it('gap beyond window resets mult to base', () => {
+  it('gap ≥500ms (idle) resets mult to base', () => {
     const s = initWheelAccel(false, 1)
 
     for (let t = 1000; t < 1100; t += 20) {
       computeWheelStep(s, 1, t)
     }
 
+    // 900ms gap — well past the 500ms sustained window → hard reset.
     expect(computeWheelStep(s, 1, 2000)).toBe(1)
+  })
+
+  // Regression for #98295: modern terminals (Windows Terminal, Ghostty,
+  // iTerm2, Alacritty) quantise touchpad deltas and emit SGR events at
+  // ~80-300ms intervals — well above the old 40ms fast-burst window.
+  // Before the fix, every such event would hard-reset mult to base (1),
+  // so acceleration never engaged.
+  it('sustained scroll at ~120ms intervals ramps above base (#98295)', () => {
+    const s = initWheelAccel(false, 1)
+
+    // Simulate 10 scroll events at 120ms cadence (typical Windows Terminal).
+    computeWheelStep(s, 1, 1000)
+    for (let i = 1; i <= 9; i++) {
+      computeWheelStep(s, 1, 1000 + i * 120)
+    }
+
+    // After 9 sustained same-direction events the multiplier must be > 1.
+    expect(s.mult).toBeGreaterThan(1)
+  })
+
+  it('sustained scroll stays within cap', () => {
+    const s = initWheelAccel(false, 1)
+
+    // 30 events at 100ms each — multiplier must never exceed WHEEL_ACCEL_MAX (6).
+    computeWheelStep(s, 1, 1000)
+    for (let i = 1; i <= 30; i++) {
+      const rows = computeWheelStep(s, 1, 1000 + i * 100)
+      expect(rows).toBeLessThanOrEqual(6)
+    }
   })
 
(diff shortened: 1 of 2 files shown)


alt-glitch:

Related to #83675: both target #98295, but this PR uses a sustained half-life ramp while #83675 uses bounded native fast-ramp and decay bands. These are competing implementations for maintainer selection.

PR #84981, opened 2026-08-13
fix(desktop): venv-blocker scan timeout aborts every in-app update on managed Windows hosts

baihemax:
## Problem

In-app updates (the Desktop updater) fail on every attempt with:

```
[updates] venv-blocker probe failed: exit code -1
[updates] error: Update aborted: Desktop could not verify the Hermes installation is free.
```

"exit code -1" is the diagnostic rendering of a **timeout kill**: `execFile` kills the scan subprocess at 15s, leaving `status`/`code` null, which `err.status ?? err.code ?? -1` renders as -1.

## Root cause — two compounding problems

1. **The scan is genuinely slow on managed Windows hosts.** `_detect_venv_python_processes` prefetches the `cwd` attribute for **every process** via `psutil.process_iter([... , "cwd"])`. On Windows, psutil resolves cwd by opening the target process and reading its PEB — AV/EDR hooks can stall that for seconds per process. Measured on a managed 450-process box: **53s** for a full scan. The 15s `SCAN_TIMEOUT_MS` in `venv-blocker-scan.ts` kills it every time.
2. **A failed probe aborts the handoff with no retry.** `main.ts` treats `probe-failure` as fatal, so even a transient AV spike fails the update permanently for that attempt.

## Fix

- **`hermes_cli/update_cmd.py`**: stop prefetching `cwd` for every process; query `proc.cwd()` on demand only for the few candidate processes that reach the `hermes_cli.main`-without-root-in-cmdline fallback. Scan drops from 53s to ~13-16s.
- **`apps/desktop/electron/venv-blocker-scan.ts`**: raise `SCAN_TIMEOUT_MS` 15s → 60s (slowest observed scan + margin), and retry the scan **once** on probe-failure before giving up, so a transient stall no longer fails the update. Deterministic failures (venv python missing) skip the retry.

## Tests

- Electron: 4 new cases in `venv-blocker-scan.test.ts` (retry-once recovers, both-attempts-fail surfaces probe-failure, missing python does not retry, timeout >= 60s). 23/23 pass.
- Python: existing `test_update_venv_health.py` + `test_scan_venv_blockers.py` still pass (29/29).

PR #75570, opened 2026-07-31
fix(desktop): prevent venv scan timeout on busy Windows hosts

andyst-dev:
## Summary

- avoid eager `cmdline` and `cwd` queries for every Windows process by fetching them only for plausible Python/uv/Hermes candidates
- preserve direct-venv and external-interpreter blocker detection while skipping expensive fallback inspection for unrelated processes
- raise the Desktop watchdog from 15s to 60s and identify timeout failures explicitly instead of presenting them like a confirmed blocker

Fixes #75460.

## Why

The blocker scan currently asks psutil to prefetch `cmdline` and `cwd` for the entire process table. Those are expensive per-process Windows queries; on the reported 574-process host, the subprocess reaches Electron's 15-second watchdog and is killed before returning a result.

The first pass now requests only `pid`, `exe`, and `name`. `cmdline` is queried lazily for processes whose executable is already in the managed venv or whose name can plausibly be a Python/uv/Hermes trampoline. `cwd` is queried only for the install-root fallback. The conservative fallback behavior remains covered by regression tests.

The watchdog remains as a safety boundary, but 60 seconds leaves headroom for unusually busy Windows hosts. If it is reached, Desktop now says that the verification scan timed out and that no blocking process was confirmed.

## Testing

- `npm run test:desktop:platforms` — 901 passed, 2 skipped
- `python -m pytest -q tests/hermes_cli/test_update_venv_health.py tests/hermes_cli/test_scan_venv_blockers.py` — 14 passed
- Desktop typecheck — passed
- targeted ESLint and Prettier checks — passed
- Ruff and Python byte-compilation — passed
- Desktop production build — passed
- `git diff --check` — passed

The Windows process-table timing itself was reported on Windows and is represented here with injected psutil process objects; I did not claim a physical Windows reproduction.

## Related

#74831 adds a settling retry for transient `blocked` verdicts after backend shutdown. This PR addresses a distinct failure mode: the scan subprocess itself timing out before it can produce any verdict.

PR #98521, opened 2026-08-30
fix(desktop): raise venv-blocker scan timeout from 15s to 60s

### LINKED ISSUE #98377: [Bug]: venv-blocker scan exceeds its own 15s timeout on hosts with many processes — update fails with no visible reason
## Summary

On Windows hosts with a few hundred running processes, the desktop updater's
venv-blocker scan consistently exceeds its own 15-second timeout. The update
button then fails with no actionable reason surfaced in the UI, and the user
has no way to tell that a timeout — rather than a real blocker — caused it.

The probe itself is not stuck. It completes correctly; it just needs more
wall-clock time than `SCAN_TIMEOUT_MS` allows.

## Environment

- OS: Windows 11 Pro 26200
- Hermes desktop app, updating from the in-app update button
- `hermes-agent` at `%LOCALAPPDATA%\hermes\hermes-agent`
- Running processes at measurement time: **446**

## Measurement

Running the exact module the scan runs, with the exact interpreter the scan
resolves, three consecutive times:

```
$ venv/Scripts/python.exe -m hermes_cli._scan_venv_blockers
run 1: 24638 ms
run 2: 20671 ms
run 3: 22749 ms
```

The timeout is `15000 ms`:

```ts
// apps/desktop/electron/venv-blocker-scan.ts:48
const SCAN_TIMEOUT_MS = 15000
```

```ts
// apps/desktop/electron/venv-blocker-scan.ts:227-233
const proc = await execFn(venvPython, ['-m', SCAN_MODULE], {
  cwd: updateRoot,
  encoding: 'utf-8',
  timeout: SCAN_TIMEOUT_MS,
  windowsHide: true
} as any)
```

Every run exceeds the limit by 5.7 to 9.6 seconds, so the scan never returns
`clear` or `blocked` on this machine — it always lands in `probe-failure`.

## Steps to reproduce

1. Use a Windows host with roughly 400+ running processes. (A normal desktop
   with a browser, an editor, a chat client and a few background services
   reaches this easily.)
2. Open the Hermes desktop app.
3. Click the update button.

## Expected

Either the scan completes, or the failure message distinguishes "the scan
timed out" from "a blocking process was found", so the user knows whether to
close something or to retry.

## Actual

The update fails. The UI does not surface that a timeout occurred. From the
user's side the button simply does not work, repeatedly, with no path forward.

The `probe-failure` branch does build a diagnostic string:

```ts
// apps/desktop/electron/venv-blocker-scan.ts:238-246
const diag = [`exit code ${err.status ?? err.code ?? -1}`]
if (err.stderr) { diag.push(String(err.stderr).slice(0, 200)) }
return { kind: 'probe-failure', error: diag.join('; ') }
```

On a timeout there is typically no `stderr`, so the diagnostic collapses to an
exit code, which does not tell the user or a support reader that the cause was
the time limit.

## Why the probe is slow

`psutil.process_iter` has to read the command line of every process. On this
host that is 446 processes, and each `cmdline` read on Windows is a separate
API call into another process's memory. The cost scales with process count,
not with anything the user can control from the app.

## Suggested directions

Listed roughly by how much they change:

1. **Raise `SCAN_TIMEOUT_MS`.** A limit around 60 seconds would cover this
   host with margin. Cheapest fix, but it only moves the threshold.
2. **Scale the timeout with process count.** Read the count first (cheap),
   then allow proportional time.
3. **Report timeouts distinctly.** Add a `kind: 'probe-timeout'` outcome, or
   include the elapsed time and the limit in the `probe-failure` error string,
   so the message says what actually happened.
4. **Narrow what the scan reads.** The classifier only needs `cmdline` for
   processes whose name matches `python(w)?(\.exe)?`. Filtering on name before
   reading command lines would cut most of the cost.

I am happy to open a PR for whichever direction you prefer.

salch-cred:
## Problem

On Windows hosts with many running processes (400+), the psutil-based venv-blocker scan in \pps/desktop/electron/venv-blocker-scan.ts\ takes 20–25 s. The hard 15 s \SCAN_TIMEOUT_MS\ cap caused every in-app update to fail silently with a \probe-failure\ result — no actionable error was shown in the UI.

Measured in #98377:
\\\
run 1: 24 638 ms
run 2: 20 671 ms
run 3: 22 749 ms
\\\

## Fix

Raise \SCAN_TIMEOUT_MS\ from 15 000 → 60 000 ms. The scan is still bounded; 60 s covers the realistic worst case (25 s) with comfortable headroom.

Fixes #98377

### DIFF
--- a/apps/desktop/electron/venv-blocker-scan.ts
+++ b/apps/desktop/electron/venv-blocker-scan.ts
@@ -45,7 +45,13 @@ export type ScanOutcome =
 // Constants
 // ---------------------------------------------------------------------------
 
-const SCAN_TIMEOUT_MS = 15000
+// The venv-blocker scan invokes `python -m hermes_cli._scan_venv_blockers`
+// which enumerates all running processes via psutil.  On Windows hosts with
+// many running processes (400+) the scan takes 20-25 s.  The old 15 s hard
+// cap caused every update attempt to fail with a silent probe-failure on
+// such machines, with no actionable error surfaced in the UI (#98377).
+// Raised to 60 s — still bounded, but accommodates the realistic worst case.
+const SCAN_TIMEOUT_MS = 60000
 const SCAN_MODULE = 'hermes_cli._scan_venv_blockers'
 
 // ---------------------------------------------------------------------------


alt-glitch:

Related: #75570 and #84981 are open broader alternatives for the Windows venv-blocker timeout family. This PR is the focused timeout-only variant.

teknium1:
Thanks @salch-cred. This is the timeout-constant subset of the earlier #75570 (@andyst-dev, 2026-07-31), which raises the same watchdog 15s → 60s AND fixes the underlying scan cost (lazy `cmdline`/`cwd` reads gated behind a cheap name filter) plus explicit timed-out messaging. #75570 has been salvaged onto current main as #99674 (windows-latest live-lane proof: https://github.com/NousResearch/hermes-agent/actions/runs/33427285478). Raising the timeout alone would leave a 20-25s scan UX on hosts the perf fix drops to ~1-2s. Closing as a duplicate of the earlier, broader fix — credit to @andyst-dev as first submitter of the timeout family, and to you for the independent #98377 measurement confirming the same watchdog failure.

PR #33826, opened 2026-05-28
fix(agent): stop probe stepdown on context overflow without provider limit (salvage #33673)

teknium1:
## Summary
A configured `context_length` no longer silently shrinks when a provider returns a generic context-overflow error with no concrete max. Compression handles the long-conversation case; the configured window is preserved.

Root cause: when `parse_context_limit_from_error()` couldn't extract a number, the overflow handler fell back to `get_next_probe_tier(old_ctx)` — turning a 1M session into 256K → 128K → 64K on repeated overflows with no config change.

## Changes
- `agent/model_metadata.py`: new `get_context_length_from_provider_error()` returning a provider-reported lower limit or `None`.
- `agent/conversation_loop.py`: drop the `get_next_probe_tier()` fallback in the overflow recovery path. Keep `context_length` and compress when no provider limit is reported; still use the parsed limit when present; preserve the Minimax delta-only branch.
- `tests/test_ctx_halving_fix.py`: `TestContextOverflowLimitSelection` covers generic overflow without a limit, explicit provider limit, and reported-limit ≥ current.

## Validation
| | Result |
|---|---|
| tests/test_ctx_halving_fix.py | 27 passed |
| tests/agent/test_model_metadata.py | 102 passed |
| tests/agent/test_context_compressor.py | 83 passed |
| tests/run_agent/test_1630_context_overflow_loop.py | 16 passed |
| tests/run_agent/test_long_context_tier_429.py | 15 passed |

Salvage of #33673 by @yangguangjin onto current main. Authorship preserved via rebase-merge. Fixes #33669, supersedes the narrower #14953 by @atmigtnca (credit retained).

## Infographic

context-overflow-trust-provider

Issue #33669, opened 2026-05-28
Context overflow without provider max should not probe-step context length


yangguangjin:
## Bug Description

When a provider returns a context-overflow error but does not include the actual maximum context length, Hermes currently falls back to the next built-in probe tier. For a session configured with a 1M context window, that means the running agent can silently change its effective context length from `1,000,000` to `256,000`, then continue descending to `128,000`, `64,000`, etc. on repeated overflow retries.

This makes a configured long-context session appear to lose its 1M window even though the config file is unchanged.

Related to #32423, but this issue tracks the specific root cause in context-overflow recovery: guessed probe-tier step-down when the provider did not report a concrete max context limit.

## Steps to Reproduce

1. Configure a provider/model with a large context length, for example:

```yaml
model:
  provider: custom
  default: gpt-5.5
  context_length: 1000000
```

2. Run a long session until the provider rejects a request with a generic context overflow error such as:

```text
Your input exceeds the context window of this model. Please adjust your input and try again.
```

3. Ensure the provider error text does not include a concrete maximum like `272000 maximum` or `maximum context length is 272000`.
4. Observe Hermes' recovery path.

## Expected Behavior

If the provider does not report a concrete max context length, Hermes should:

- keep the configured/runtime `context_length` unchanged;
- attempt conversation compression;
- retry only if compression actually reduced the message set;
- avoid treating built-in probe tiers as provider-confirmed context limits.

If the provider does report a smaller concrete limit, Hermes may use that provider-reported limit.

## Actual Behavior

Hermes calls the probe-tier fallback and logs a step-down similar to:

```text
Context length exceeded — stepping down: 1,000,000 → 256,000 tokens
Context too large (...) — compressing (...)
```

If compression aborts or does not reduce the payload enough, subsequent overflow attempts can continue stepping down through lower probe tiers.

## Proposed Fix

- Add a helper that only returns provider-reported lower context limits.
- In the conversation loop, stop using `get_next_probe_tier()` as a fallback when `parse_context_limit_from_error()` cannot extract a concrete max.
- Keep context length unchanged and rely on compression when the provider gives no max.
- Continue allowing provider-confirmed lower limits to update the compressor context length.

## Test Coverage

Add regression tests for:

- generic overflow without a provider limit returns no context-length update;
- explicit provider limit still updates to that limit;
- provider limit greater than or equal to the current runtime context is ignored.

yangguangjin:
[...]

PR #98522, opened 2026-08-30
fix(bedrock): stop the context probe escalating on rejections it cannot parse

### LINKED ISSUE #98520: Bedrock context probe escalates on the wrong branch: tier 2 uploads 12.2 MB that can never return new information
## Bug Description

`probe_bedrock_context_length()` (`agent/bedrock_adapter.py`) walks `_BEDROCK_PROBE_TIERS = (1_300_000, 2_200_000)`, padding a request past each tier to read the real window out of Bedrock's length-validation error. The loop escalates to the **larger** tier when a request is **rejected**, and `return`s immediately when a request is **accepted**.

Both branches are the wrong way round, and the consequence is that **tier 2 can never produce information tier 1 didn't** — it is a wasted 12.2 MB upload in every case it is reached.

The reason is that the error shape and the reported `maximum` are properties of the **model**, not of the payload size. Bedrock has two length-error shapes:

```
prompt is too long: 1444476 tokens > 1000000 maximum     ← numeric, parseable
Input is too long for requested model.                   ← numberless, unparseable
```

Which one you get depends on the model. So if tier 1 gave a numberless error, tier 2 gives the same numberless error; if tier 1 gave a number, tier 2 gives the same number. Padding harder cannot change either.

Meanwhile the docstring's stated reason for having tiers at all —

> (b) stepping up discovers larger windows without over-padding smaller ones

— is never actually exercised, because acceptance `return`s instead of escalating. A model that accepts a 1.3M-token prompt returns `1_300_000` and never tries 2.2M. **Tier 2 is dead code for its documented purpose, and live only on the path where it cannot help.**

The in-code comment already half-notices the contradiction:

```python
# No parseable limit at this tier (opaque server error, auth,
# throttle).  Try the next, smaller-overage strategy is N/A here —
# tiers ascend — so just continue; if all fail we return None.
continue
```

This is the same principle #33669 established for the *recovery* path ("if the provider does not report a concrete max, don't guess with probe tiers"), fixed in #33826. That fix landed in `agent/conversation_loop.py` and `agent/model_metadata.py`; the **discovery** path in `agent/bedrock_adapter.py` was not part of it and still guesses by escalating.

## Steps to Reproduce

Any Bedrock model that returns the numberless shape — `claude-sonnet-4` and `claude-sonnet-4-5` both do:

```python
from agent.bedrock_adapter import probe_bedrock_context_length
probe_bedrock_context_length("us.anthropic.claude-sonnet-4-5-20250929-v1:0", "us-east-2")
```

Measured live in `us-east-2` on 2026-08-30 (`boto3` 1.42.89):

| model | tier | bytes uploaded | elapsed | error code | reported max | `x-amzn-RequestId` |
|---|---|---|---|---|---|---|
| `claude-sonnet-4-5` | 1 (1.3M) | 7,222,220 | 0.50s | ValidationException | *(none)* | `58d9ac48-48a0-4fa7-a513-1043c47105c0` |
| `claude-sonnet-4-5` | 2 (2.2M) | 12,222,220 | 0.82s | ValidationException | *(none)* | `4569eaad-2ade-4c4c-8c94-6809206b811e` |
| `claude-sonnet-4` | 1 (1.3M) | 7,222,220 | 0.62s | ValidationException | *(none)* | `9c15f990-3ba3-4e47-a8c2-c89940b4cd9c` |
| `claude-sonnet-4` | 2 (2.2M) | 12,222,220 | 0.74s | ValidationException | *(none)* | `8414a6bf-be48-4027-84e5-ee72ef1de43c` |
| `claude-opus-4-8` | 1 (1.3M) | 7,222,220 | 2.63s | ValidationException | `1000000` | `8b984dbf-62df-42b2-a224-2b129f26418f` |
| `claude-opus-4-8` | 2 (2.2M) | 12,222,220 | 3.65s | ValidationException | `1000000` | `184c3f83-3063-435a-be48-2fe7b2d160e0` |

Two things the `opus-4-8` rows show: tier 2 reports **the same** `1000000` as tier 1, so escalation adds nothing even in the parseable case — and production never reaches tier 2 there anyway, because tier 1 returns.

`RetryAttempts=0` on every call, so there is no boto3 retry amplification — the 12.2 MB is uploaded exactly once.

## Expected Behavior

- A rejection carrying a parseable `maximum` → return it. (Already correct.)
- A rejection **without** a parseable number → stop. The next tier will produce the identical unparseable error; return `None` and let the caller fall back to the static table.
- An **acceptance** → *this* is the case where a larger tier is informative, because the window is only known to be "at least this". Escalate, and keep the accepted tier as a floor in case the larger request fails.

## Actual Behavior

Rejection escalates and acceptance returns — so `claude-sonnet-4-5` uploads 7.2 MB, learns nothing, then uploads a further **12.2 MB** to learn the same nothing, before returning `None`.

## Impact

Moderate, not severe — the final answer is still correct (`None` → static table → 200,000, which is right), and `get_bedrock_context_length()` caches per model, so the cost is paid once rather than every turn. What it costs is a first-use stall: in-region the wasted tier is only ~0.8s, but the 12.2 MB body is uploaded from the *user's* machine, so on a 10 Mbps uplink it is roughly 10 seconds of a startup hang with nothing on screen explaining it — and it is pure waste, since the request cannot return new information.

## Proposed Fix

Swap the two branches so escalation happens on the branch where it is informative:

- rejection + parseable limit → `return limit` (unchanged)
- rejection + no parseable limit → `break`, then `return None`
- acceptance → record the tier as a floor and `continue` to the next tier; return the best accepted tier after the loop

This makes tier 2 reachable for the purpose the docstring claims for it, deletes the 12.2 MB dead upload, and keeps `None` as the contract when nothing is parseable. It also preserves the docstring's reason (a): if an escalated request fails opaquely because it is *wildly* oversized, the already-accepted lower tier is still returned rather than lost.

I have this implemented with unit tests for all five paths and can open the PR — happy to hold if a maintainer would rather shape it differently first.

JiaDe-Wu:
Fixes #98520.

## What

`probe_bedrock_context_length()` escalates to the **larger** probe tier when a request is **rejected**, and returns immediately when one is **accepted**. Both branches are the wrong way round, and the consequence is that **tier 2 can never return information tier 1 didn't** — it is a 12.2 MB upload that is guaranteed to be wasted in every case it is reached.

A rejection is terminal, because the error shape follows the **model**, not the payload size. Bedrock has two shapes:

```
prompt is too long: 1444476 tokens > 1000000 maximum     ← parseable
Input is too long for requested model.                   ← no number
```

Padding harder never converts the second into the first. `claude-sonnet-4` and `claude-sonnet-4-5` return the numberless shape at *every* tier, so today they upload 7.2 MB, learn nothing, then upload a further **12.2 MB** to learn the same nothing before returning `None`.

Meanwhile the docstring's stated reason for having tiers at all —

> (b) stepping up discovers larger windows without over-padding smaller ones

— was never exercised, because acceptance `return`s instead of escalating. A model that accepts a 1.3M-token prompt returned `1_300_000` and never tried 2.2M. **Tier 2 was dead code for its documented purpose, and live only on the path where it cannot help.**

The in-code comment already half-noticed this:

```python
# No parseable limit at this tier (opaque server error, auth,
# throttle).  Try the next, smaller-overage strategy is N/A here —
# tiers ascend — so just continue; if all fail we return None.
continue
```

## The change

Swap which branch escalates:

| outcome | before | after |
|---|---|---|
| rejected, parseable `maximum` | return it | return it *(unchanged)* |
| rejected, nothing parseable | **escalate** | **stop**, return `None` |
| accepted | **return the tier** | **escalate**, keeping the tier as a floor |

Acceptance is the only outcome that leaves the window unbounded above, so it is the only one a larger tier can improve on. The floor matters for the hazard the docstring's reason (a) describes: if the escalated request fails opaquely because it is *wildly* oversized, the bound already confirmed by the accepted tier is returned rather than discarded.

`agent/bedrock_adapter.py` is +27/−8, all inside this one function plus its docstring. No config, no new env var, no signature or contract change — `None` still means "couldn't determine, fall back to the static table".

## Verified against live Bedrock

`us-east-2`, 2026-08-30, `boto3` 1.42.89 — A/B through the **production function**, counting real requests and bytes on the wire:

| model | error shape | before | after |
|---|---|---|---|
| `us.anthropic.claude-sonnet-4-5-20250929-v1:0` | numberless | `None`, **2** requests, 19.4 MB | `None`, **1** request, **7.2 MB** |
| `us.anthropic.claude-sonnet-4-20250514-v1:0` | numberless | `None`, **2** requests, 19.4 MB | `None`, **1** request, **7.2 MB** |
| `us.anthropic.claude-opus-4-8` | numeric | `1000000`, 1 request, 7.2 MB | `1000000`, 1 request, 7.2 MB |

Identical results, 12.2 MB less uploaded for each affected model. Raw per-tier evidence with `x-amzn-RequestId`s for both shapes is in #98520 — including `claude-opus-4-8` returning the *same* `1000000` at tier 2 as at tier 1, which is the same point from the other direction: escalation adds nothing even when the error is parseable.

`RetryAttempts=0` on every call, so there is no boto3 retry amplification — the wasted 12.2 MB is uploaded exactly once, not per retry.

## Impact, stated fairly

This is a waste-and-clarity fix, not a correctness fix. The final answer was already right: `None` → static table → 200,000 for both sonnet models, which is correct, and `get_bedrock_context_length()` caches per model so the cost is paid once rather than every turn.

What it costs is a first-use stall. In-region the wasted tier is only ~0.8s, but the body is uploaded from the **user's** machine — on a 10 Mbps uplink 12.2 MB is roughly 10 seconds of startup hang with nothing on screen explaining it. The fix also makes tier 2 reachable for the purpose it was written for, which matters the first time AWS ships a window above 1.3M.

## Tests

`TestBedrockContextProbeTierEscalation`, 7 cases in `tests/agent/test_bedrock_adapter.py`, asserting `converse.call_count` — the point is not just the return value but that the second request is **not sent**. The numberless-error fixture is the verbatim string from the live sweep, with the RequestIds in a comment so a future reader can tell it is real rather than invented.

Honestly labelled: **5 of the 7 fail on unpatched `main`**; the other two (`test_parseable_rejection_returns_on_the_first_tier`, `test_tiers_ascend`) assert behaviour that already held and are there to pin it down, not as regression tests.

```
FAILED ...::test_numberless_rejection_does_not_escalate
FAILED ...::test_opaque_failure_does_not_escalate
FAILED ...::test_acceptance_escalates_to_bound_the_window
FAILED ...::test_accepted_floor_survives_a_useless_next_tier
FAILED ...::test_every_tier_accepted_returns_the_largest
5 failed, 2 passed
```

with e.g.

```
AssertionError: probe escalated to a larger tier after a rejection it could not
parse; the larger request returns the identical error, so this is a wasted
multi-megabyte upload
assert 2 == 1
```

## Relation to #33826 (merged)

#33669 established exactly this principle for the context-overflow **recovery** path — *"if the provider does not report a concrete max, don't guess with probe tiers"* — and #33826 implemented it in `agent/conversation_loop.py` and `agent/model_metadata.py`. The **discovery** path in `agent/bedrock_adapter.py` was not part of that change and still guessed, in the opposite direction. Different function, different module, no overlap; `tests/test_ctx_halving_fix.py` (that fix's regression test) is in the run below and still passes.

## Testing

- `pytest tests/agent/test_bedrock_adapter.py tests/agent/test_model_metadata.py tests/agent/test_bedrock_1m_context.py tests/agent/test_bedrock_empty_text_blocks.py tests/test_ctx_halving_fix.py` → **213 passed**
- new class alone → 7 passed; against unpatched `agent/bedrock_adapter.py` → 5 failed, 2 passed (above)
- `ruff check` → clean
- live A/B against real Bedrock → table above

### DIFF
--- a/agent/bedrock_adapter.py
+++ b/agent/bedrock_adapter.py
@@ -1868,6 +1868,16 @@ def probe_bedrock_context_length(model_id: str, region: str) -> Optional[int]:
     InternalServerException instead of a clean length error, and (b) stepping
     up discovers larger windows without over-padding smaller ones.
 
+    Escalation therefore only happens on **acceptance**, which is the one
+    outcome that leaves the window unbounded above.  A *rejection* is terminal:
+    the error shape is a property of the model, not of the payload size, so a
+    bigger prompt returns the identical error.  Bedrock has two shapes —
+
+        prompt is too long: 1444476 tokens > 1000000 maximum   (parseable)
+        Input is too long for requested model.                 (no number)
+
+    — and padding harder never converts the second into the first.
+
     Returns the detected window, or ``None`` if the probe could not run
     (missing credentials, network error, or no parseable limit) so the caller
     can fall back to the static table.
@@ -1884,6 +1894,7 @@ def probe_bedrock_context_length(model_id: str, region: str) -> Optional[int]:
         return None
 
     last_error = ""
+    largest_accepted = None
     for tier_tokens in _BEDROCK_PROBE_TIERS:
         pad_words = int(tier_tokens / _WORDS_PER_TOKEN)
         oversized = "data " * pad_words
@@ -1893,14 +1904,15 @@ def probe_bedrock_context_length(model_id: str, region: str) -> Optional[int]:
                 messages=[{"role": "user", "content": [{"text": oversized}]}],
                 inferenceConfig={"maxTokens": 8},
             )
-            # Accepted a prompt this large → the window is at least this tier.
-            # Returning the tier as a lower bound is safe and avoids inventing
-            # a number we can't confirm.
+            # Accepted a prompt this large → the window is at least this tier,
+            # but we have no upper bound yet.  This is the only outcome a bigger
+            # tier can improve on, so keep this as a floor and step up.
             logger.debug(
                 "Bedrock context probe for %s accepted ~%s-token prompt; "
                 "window is at least that", model_id, f"{tier_tokens:,}",
             )
-            return tier_tokens
+            largest_accepted = tier_tokens
+            continue
         except Exception as exc:
             msg = str(exc)
             last_error = msg
@@ -1911,10 +1923,17 @@ def probe_bedrock_context_length(model_id: str, region: str) -> Optional[int]:
                     model_id, f"{limit:,}",
                 )
                 return limit
-            # No parseable limit at this tier (opaque server error, auth,
-            # throttle).  Try the next, smaller-overage strategy is N/A here —
-            # tiers ascend — so just continue; if all fail we return None.
-            continue
+            # Rejected with nothing parseable — a numberless length error, or an
+            # opaque server/auth/throttle failure.  Escalating cannot help: the
+            # error shape follows the model, not the payload size, and per (a)
+            # above a larger payload is *more* likely to fail opaquely.  Stop
+            # rather than upload the next tier to learn the same nothing.
+            break
+
+    if largest_accepted is not None:
+        # An earlier tier was accepted and a later one gave us nothing better;
+        # the confirmed lower bound still beats falling back to the table.
+        return largest_accepted
 
     logger.debug(
         "Bedrock context probe for %s returned no parseable limit: %s",
--- a/tests/agent/test_bedrock_adapter.py
+++ b/tests/agent/test_bedrock_adapter.py
@@ -1012,6 +1012,125 @@ class TestBedrockContextProbe:
                 region="eu-central-1") == 1_000_000
 
 
+class TestBedrockContextProbeTierEscalation:
+    """Which outcome is allowed to step up to the next probe tier.
+
+    Escalation is only informative after an *acceptance*, because that is the
+    only outcome that leaves the window unbounded above.  A rejection is
+    terminal: the error shape follows the model, not the payload size, so a
+    bigger prompt returns the identical error.  Each tier is a multi-megabyte
+    upload from the user's machine (~7.2 MB then ~12.2 MB at the current tiers),
+    so a tier that cannot return new information must not be sent.
+    """
+
+    # Verbatim from live Bedrock, us-east-2, 2026-08-30.  claude-sonnet-4 and
+    # claude-sonnet-4-5 both return this shape, with no number to parse, at
+    # *every* tier (RequestIds 58d9ac48-48a0-4fa7-a513-1043c47105c0 at 1.3M and
+    # 4569eaad-2ade-4c4c-8c94-6809206b811e at 2.2M).
+    NUMBERLESS_LENGTH_ERROR = (
+        "An error occurred (ValidationException) when calling the Converse "
+        "operation: The model returned the following errors: Input is too long "
+        "for requested model."
+    )
+    # The other shape, from the same sweep (claude-opus-4-8, RequestId
+    # 8b984dbf-62df-42b2-a224-2b129f26418f).
+    NUMERIC_LENGTH_ERROR = (
+        "An error occurred (ValidationException) when calling the Converse "
+        "operation: The model returned the following errors: prompt is too "
+        "long: 1444476 tokens > 1000000 maximum"
+    )
+
+    def _client(self, *outcomes):
+        """A client whose successive converse() calls follow ``outcomes``.
+
+        An ``Exception`` instance is raised; anything else is returned.
+        """
+        client = MagicMock()
+        client.converse.side_effect = list(outcomes)
+        return client
+
+    def _probe(self, client):
+        from agent.bedrock_adapter import probe_bedrock_context_length
+        with patch("agent.bedrock_adapter._get_bedrock_runtime_client",
+                   return_value=client):
+            return probe_bedrock_context_length("any.model", "us-east-2")
+
+    def test_numberless_rejection_does_not_escalate(self):
+        """The regression this class exists for.
+
+        A numberless length error means *this model* does not report its
+        maximum.  The next tier returns the same numberless error, so sending it
+        uploads ~12.2 MB to learn exactly nothing.
+        """
+        client = self._client(
+            Exception(self.NUMBERLESS_LENGTH_ERROR),
+            Exception(self.NUMBERLESS_LENGTH_ERROR),
+        )
+        assert self._probe(client) is None
+        assert client.converse.call_count == 1, (
+            "probe escalated to a larger tier after a rejection it could not "
+            "parse; the larger request returns the identical error, so this is "
+            "a wasted multi-megabyte upload"
+        )
+
+    def test_opaque_failure_does_not_escalate(self):
+        """Same for a non-length failure.
+
+        Auth and throttle errors recur regardless of size, and per the
+        function's own docstring a *larger* payload is more likely to produce an
+        opaque InternalServerException — so stepping up is strictly worse.
+        """
+        client = self._client(
+            Exception("An error occurred (InternalServerException) ..."),
+            Exception("An error occurred (InternalServerException) ..."),
+        )
+        assert self._probe(client) is None
+        assert client.converse.call_count == 1
+
+    def test_parseable_rejection_returns_on_the_first_tier(self):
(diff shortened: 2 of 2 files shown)


Enough1122:

### PR #98522 — fix(bedrock): stop the context probe escalating on rejections it cannot parse

**Summary:** Corrects inverted escalation in `probe_bedrock_context_length`: rejection (parseable or numberless) is now terminal (`break`), acceptance escalates and records `largest_accepted` floor. Prevents wasted 12.2 MB tier-2 upload when tier-1 already returned numberless `Input is too long` error (identical across tiers per live sweeps). Adds `largest_accepted` return when later tier fails opaquely.

- Docstring now explains two Bedrock error shapes (numeric `tokens > max` vs numberless) and why bigger payload cannot convert second into first — accurate per RequestIds cited in tests.
- `test_numberless_rejection_does_not_escalate` pins `call_count==1` — regression guard for the 12 MB waste.
- `test_accepted_floor_survives_a_useless_next_tier` ensures we don't discard confirmed lower bound — beats falling back to static table.
- Tiers remain ascending (pinned by `test_tiers_ascend`) so floor logic is valid.

**Verdict:** Correct, well-evidenced fix with thorough tests. No blocking issues.

JiaDe-Wu:
Closing this in favour of a rewrite against current `main`: **#106632**.

The mechanical reason is staleness — this branch is 6,737 commits behind and `agent/bedrock_adapter.py` has been touched 21 times over its base, so it does not merge and the probe it patches has since been rewritten into a much shorter form. It needed rebuilding rather than rebasing.

But I re-measured against live Converse before replacing it, and I would rather say the second reason plainly than let the diff carry it forward.

**The docstring I added here is wrong.** It says:

> the error shape is a property of the model, not of the payload size, so a bigger prompt returns the identical error

Bedrock does not behave that way. Same model, same region, only the padding differs:

| target tokens | payload | outcome |
|---|---|---|
| 2,200,000 | 12.2 MB | `This model's maximum context length is 1048576 tokens. Please reduce the length of the prompt` |
| 5,000,000 | 27.8 MB | `Input is too long for requested model.` — no number at all |

The shape **does** follow payload size, and it gets *worse* as the payload grows. So the `break` this PR introduces is the right behaviour, but the reason given for it is false — and a reviewer who accepted this would have been accepting the false premise along with it. The replacement argues it the other way round: escalation is what *produces* the numberless shape, which is the argument against escalating into it.

Two more things I got wrong here, both found by measuring:

- **I could not reproduce the numberless error inside the actual tiers.** `_BEDROCK_PROBE_TIERS` tops out at 2,200,000, and at that size rejections are still parseable — `us.meta.llama3-3-70b-instruct-v1:0` returns `maximum context length is 131072 tokens` at both 1.3M and 2.2M, and Llama 4 Maverick returns `1048576` at 2.2M. The numberless shape needed a 5,000,000 target to appear. So the in-tier failure this fix actually needs to handle is not the one I described.
- **The in-tier failure that is real is a throttle.** Llama 4 Scout at the 2.2M tier came back `ThrottlingException … Too many tokens, please wait before trying again.` after 256 s, with boto3 having already spent its four internal retries. That raises exactly like a length rejection and carries no window. Escalating after it uploads a larger prompt to a service that has just asked you to wait.

Finally, the `largest_accepted` half of this PR does not fix the acceptance bug — it moves it. Returning the largest accepted tier still reports a floor as if it were the window: Scout accepts 2,222,258 tokens against a real 3,500,000 window, so this PR would answer `2,200,000` for a 3.5M model instead of the current `1,300,000`. Better, still wrong, and `agent/model_metadata.py` persists probe results, so the truncated number sticks. #106632 compares the floor against the static table before reporting it, and defers to the table when the table knows more.

#106632 also uses `agent/error_classifier.py` rather than the `THROTTLE_PATTERNS` / `classify_bedrock_error` in `bedrock_adapter.py`, since `COMPAT_MANIFEST.md:178-184` lists those as revert-scheduled plugin shims.

Nothing needed from anyone; closing it myself.

PR #98523, opened 2026-08-30
fix(tui-gateway): suppress sessions.changed broadcast on heartbeat-only mtime churn

### LINKED ISSUE #98005: sessions.changed broadcast on idle db-mtime churn (gateway_heartbeats) re-mounts chat and yanks scroll — #38015 renderer fix incomplete
## Summary

Issue #38015 was closed as fixed by PR #40433 (renderer-side: `session-states.ts` no longer republishes a content-identical transcript on `session.info` heartbeats). That fixed the **renderer** half. The **gateway** half is still live: `sessions.changed` is broadcast on idle `state.db` mtime changes that represent no session change, and the desktop still re-renders the session list + re-mounts the chat on every such broadcast — scrolling the transcript back to bottom.

## Root cause (traced, not guessed)

`tui_gateway/server.py`:

- `_CHANGE_WATCHES["sessions.changed"] = (0.5, _sessions_sig, lambda: {})` — re-checked every 0.5s.
- `_sessions_sig()` (line 4899) returns `max(st_mtime_ns of state.db, state.db-wal)`.
- `_CHANGE_BROADCAST_FLOOR_S["sessions.changed"] = 2.0`.

The shared `state.db` is written by **many profile backends + the gateway**. The cross-backend heartbeat (#94895) refreshes a row in `gateway_heartbeats` every `_HEARTBEAT_REFRESH_S = 60.0` (server.py:2050), which moves `state.db`'s mtime even when the user is idle. That flips `_sessions_sig()` → gateway broadcasts `sessions.changed` (empty payload) → desktop re-renders → chat scroll yanked to bottom.

### Evidence

Observed `sessions.changed` cadence in `desktop.log` (idle, no tool calls):

```
[2026-08-29T11:52:21.210Z] sessions.changed session_id="" payload={}
[2026-08-29T11:53:21.295Z] ...
[2026-08-29T11:54:21.382Z] ...
[2026-08-29T11:55:20.968Z] ...
[2026-08-29T11:56:21.056Z] ...
[2026-08-29T11:57:21.142Z] ...
```

Exactly +60s, every beat, empty payload — the heartbeat signature, not a real session change. `gateway_heartbeats` has 9 rows in the shared db. Environment: 28 profiles sharing one `state.db` (1.5 GB).

## Why this reproduces #38015's symptom

The renderer fix guards against `session.info` republish, but `sessions.changed` from the gateway is a *separate* event path. The desktop's session-list query + chat remount still key off it, so the same scroll-reset symptom persists. #98015 was closed as "implemented on main" but the gateway broadcast path was never addressed.

## Proposed fix

Make `_sessions_sig()` detect an **actual session change**, not any byte written to `state.db`. Minimal: key off `MAX(last_activity_at)` (or a sessions-change marker) read on a read-only connection, instead of raw db/WAL mtime. The heartbeat table and WAL checkpoints should not flip the signature.

## Workaround (local, reversible)

A local source patch changing `_sessions_sig()` to read `SELECT MAX(last_activity_at) FROM sessions` on a read-only connection eliminates the yank. Reverted via `git checkout -- tui_gateway/server.py`. Reported here so the next release carries the real fix instead of the stopgap.

## Environment

- Hermes 0.20.6 (build 2026-08-29), includes PR #40433 (`c50fb560ef` confirmed in HEAD)
- Native Wayland Linux (also likely macOS — same shared-db path)
- 28 profile backends sharing one `state.db`

Relates to #38015, #37549.

salch-cred:
## Problem

\_sessions_sig()\ in \	ui_gateway/server.py\ returns the newest mtime across \state.db\ and its WAL. \gateway_heartbeats\ refreshes a row every 60 s, bumping \state.db\'s mtime even when **no session changed**. This caused \sessions.changed\ to broadcast every ~60 s, re-mounting the Desktop chat component and yanking the scroll position.

PR #40433 fixed the *renderer* half (no re-render on content-identical \session.info\ heartbeats) but left the *gateway* half open — #98005.

## Fix

Extend \_sessions_sig\ to return a tuple of:
- \mtime_sig\ — max of \state.db\ / \state.db-wal\ mtime (fast, catches any write)
- \content_sig\ — \(session_count, max_updated_at)\ from a lightweight read-only aggregate query

Two signatures differ only when the sessions table actually changes, not on heartbeat-only mtime churn. The SQL probe fails closed to \(mtime, None)\ when the DB is unreadable, preserving the original startup/read-only behaviour.

Fixes #98005

### DIFF
--- a/tui_gateway/server.py
+++ b/tui_gateway/server.py
@@ -4898,19 +4898,47 @@ def _cron_sig():
 
 
 def _sessions_sig():
-    """Newest mtime across state.db and its WAL — the cross-process change
-    signal. Messaging-gateway turns and cron runs are written by OTHER
-    processes that never touch this gateway's transports; the shared SQLite
-    file is the one thing they all move (#58671)."""
+    """Content-based signature for the sessions table.
+
+    The cross-process change signal is a tuple of:
+    - max(state.db mtime, state.db-wal mtime)  — fast, catches any write
+    - (session_count, max_updated_at)           — content hash, filters out
+      heartbeat-only mtime churn that moves state.db without changing any
+      session row (#98005 / #38015).
+
+    Messaging-gateway turns and cron runs are written by OTHER processes that
+    never touch this gateway's transports; the shared SQLite file is the one
+    thing they all move (#58671).  But gateway_heartbeats refreshes a row
+    every 60 s, bumping state.db's mtime even when no session changed.  The
+    content component prevents that from broadcasting sessions.changed and
+    re-mounting the chat, which yanks the scroll position.
+    """
     home = _watcher_home()
-    sig = None
+    mtime_sig = None
     for name in ("state.db", "state.db-wal"):
         try:
             mtime = (home / name).stat().st_mtime_ns
         except OSError:
             continue
-        sig = mtime if sig is None else max(sig, mtime)
-    return sig
+        mtime_sig = mtime if mtime_sig is None else max(mtime_sig, mtime)
+
+    # Add a lightweight content probe: session count + newest updated_at.
+    # A failed probe falls back to mtime-only (pre-fix behaviour) so startup
+    # and read-only deployments still work.
+    content_sig = None
+    try:
+        import sqlite3 as _sqlite3
+        db_path = home / "state.db"
+        with _sqlite3.connect(f"file:{db_path}?mode=ro", uri=True, timeout=0.5) as _con:
+            row = _con.execute(
+                "SELECT count(*), max(updated_at) FROM sessions"
+            ).fetchone()
+        if row:
+            content_sig = row
+    except Exception:
+        pass
+
+    return (mtime_sig, content_sig)
 
 
 def _platforms_sig():


Enough1122:

### PR #98523 — fix(tui-gateway): suppress sessions.changed broadcast on heartbeat-only mtime churn

**Summary:** `_sessions_sig()` now returns `(mtime_sig, content_sig)` where `content_sig = (count(*), max(updated_at))` from `sessions` table via read-only SQLite query. Prevents `gateway_heartbeats` 60s mtime bumps (no session row changed) from broadcasting `sessions.changed` and remounting Desktop chat (scroll yank). Falls back to mtime-only if query fails.

- Read uses `sqlite3.connect(f"file:{db_path}?mode=ro", timeout=0.5)` — 500 ms timeout is generous for a read; failure falls back gracefully.
- Content check is `count` + `max(updated_at)` — minimal but catches any row insert/update/delete that should trigger broadcast. Edge: an update that preserves that column and count would still be suppressed, but heartbeat itself updates `updated_at` when real change occurs.
- Tuple return changes signature — callers comparing sigs still work (tuple equality), but verify no caller assumed int-only.

**Verdict:** Minimal, correct churn filter. No blocking issues.

PR #98526, opened 2026-08-30
fix(update): skip catchup fleet restart when interrupted receipt pulled SHA already matches checkout

### LINKED ISSUE #98022: [Bug]: hermes update catch-up fleet restart (#95294 fix) re-fires on every run when update_receipts/latest.json is a stale interrupted receipt — fleet restarts forever even though it is already current
# [Bug]: `hermes update` catch-up fleet restart (#95294 fix) re-fires on every run when `update_receipts/latest.json` is a stale interrupted receipt — fleet restarts forever even though it is already current

Related: #95294 (parent bug, fix landed as the catch-up restart), #91277 (umbrella), #92145 (sibling).

## Summary

The #95294 catch-up (`_apply_pending_fleet_restart_catchup`) is the right idea, but its only persistent trigger source — `logs/update_receipts/latest.json` — is never refreshed by a completed run on this install, so a single interrupted update poisons the trigger permanently. Result: **every subsequent `hermes update` prints the "did not restart running gateways" warning and restarts the whole fleet, even though the fleet is already verified current** (the same run's own fleet check prints `up to date`).

## Environment

- macOS 26.6.1, git install on `main` (fork checkout), launchd-supervised gateway, profile `default`
- Code: v0.20.6 @ `3f36c87e1e` (observed under both `f6f707b783` and `3f36c87e1e`; the relevant code is identical between them — verified with `git diff f6f707b..3f36c87e -- hermes_cli/update_cmd.py hermes_cli/update_receipt.py hermes_cli/main.py` for these paths: no changes)

## Timeline (all UTC)

| When | Event | Evidence |
|---|---|---|
| 08-28 01:28 | `hermes update` interrupted (**KeyboardInterrupt**) after the plan phase, before pull. Checkout and gateway both at `f6f707b7`. | receipt `update_20260828_093021_61623.json` (= `latest.json`): `outcome: failed`, `stop_reason: "KeyboardInterrupt: "`, `plan.runtimes[0] = {profile: default, pid: 41612, code_sha: f6f707b7, supervisor: launchd}` |
| 08-28 ~09:17 | `hermes update` succeeds: pulls 247 commits, restarts gateway, **writes no receipt file** | gateway log `Stopping gateway for restart... 2026-08-28 17:17:21` (CST); no new file in `update_receipts/` |
| 08-29 ~03:05 | `hermes update` succeeds, restarts gateway, **writes no receipt** | gateway log `2026-08-29 11:05:00` (CST); no new receipt file |
| 08-29 ~15:58 | `hermes update` pulls **115 commits** → HEAD `3f36c87e1e`, restarts gateway, fleet check prints `default (pid 60110) @ 3f36c87e — up to date`. **Writes no receipt** | user-observed CLI output |
| 08-29 ~16:0x onward | every `hermes update` (git already up to date) prints the stale warning and restarts the fleet again | user-observed CLI output; 4 consecutive runs each restarted the gateway |

As of writing, `~/.hermes/logs/update_receipts/` contains exactly 20 receipt files, the newest from 08-28 01:30 — i.e. **no receipt has been written since**, successful or not. `~/.hermes/fleet_restart_pending` does not exist.

## Why the trigger never clears (code paths, verified at `3f36c87e1e`)

1. `_pending_fleet_restart_needed()` (`hermes_cli/update_cmd.py:3021`) returns true when either the marker file `~/.hermes/fleet_restart_pending` exists **or** `_receipt_reports_stale_runtime()` is true.
2. `_receipt_reports_stale_runtime()` (`update_cmd.py:2973`) reads `latest.json`: an *unfinished* receipt (here `outcome: failed` + `stop_reason`) whose `plan.runtimes[].code_sha` lags the checkout HEAD → true. The 08-28 interrupted receipt still satisfies this, forever.
3. The up-to-date path calls `_apply_pending_fleet_restart_catchup()` (`update_cmd.py:8061` → `3195`), which on success clears **only the marker file** (`_clear_fleet_restart_pending_marker`, line 3207) — it never touches `latest.json`.
4. `latest.json` is only rewritten by `finalize_update_receipt()` (`update_receipt.py:192`), called from the success path (`update_cmd.py:9967`) / ZIP path / refusal paths / the command-boundary safety net in `cmd_update` (`main.py:10831-10869`). So a run that completes **without writing a receipt** leaves the poisoned `latest.json` in place — and the next run re-triggers the catch-up. The catch-up is idempotent, so the visible symptom is a redundant fleet restart on every update, with a warning that describes a state that no longer exists.

## The anomaly I could not fully explain (needs maintainer input)

On this machine, **no receipt is written by any completed `hermes update` since 08-28 01:30**, including 247- and 115-commit pulls that went through the full success path. The finalize call sites exist and are identical in both observed code versions. `finalize_update_receipt` swallows all write errors (`update_receipt.py:230-232`, `logger.debug` only), so a failed write is invisible.

Hypothesis (unconfirmed): the volume was near-full around then — it is currently at 93% (380/460 GiB, 33 GiB free) — and `write_text` failed (ENOSPC or similar), silently. The directory mtime of `update_receipts/` (08-28 12:42) is later than the newest file, consistent with a file-create-then-fail/delete around that window, but I cannot confirm from disk alone.

Questions:
1. Can a receipt write fail silently in a way that permanently poisons `_pending_fleet_restart_needed()`? Should the trigger be resilient to `latest.json` being stale-or-missing (e.g. bound it to the `expected_sha` written into the marker, or cross-check the live fleet before trusting an old receipt)?
2. Should the catch-up **probe the live fleet first** (`collect_fleet_versions` already reads every profile's `gateway_state.json` stamp): if every running gateway's `code_sha` equals the checkout HEAD, the obligation is already satisfied — clear the trigger instead of restarting? Today the restart is the routine, not the remedy.

## Minimal repro (logic-level)

1. Run an update that gets interrupted between `begin_update_receipt()`/plan and the pull, or after pull but before finalize → `latest.json` records an unfinished receipt with the pre-update runtime SHA.
2. Ensure no subsequent run writes a receipt (on this install that happened naturally; artificially: make `update_receipts/` unwritable after step 1, or delete the directory's write permission mid-run — the error is swallowed by design).
3. Every later `hermes update` restarts the fleet and prints the warning, even with a verified-current fleet and no marker file.

## Temporary workaround (for operators)

With a verified-current fleet (`hermes update`'s fleet matrix all `up to date`), removing the poisoned pointer stops the loop:

```
mv ~/.hermes/logs/update_receipts/latest.json ~/.hermes/logs/update_receipts/latest.json.bak
```

(Only safe when the live fleet check shows no `stale`/`down` rows — otherwise the skew is real and the restart is owed.)

## Workaround evidence on this box

Fleet is healthy: default gateway `code_sha=3f36c87e1ebdfbf7d14a88229dc9be222c12ea89` == checkout HEAD; `_pending_fleet_restart_needed()` still returns `True` (verified in-venv), so the loop will fire again on the next `hermes update` until `latest.json` is replaced by a real receipt.

salch-cred:
## Problem (#98022)

The #95294 catchup restart checks \plan.runtimes[].code_sha\ when no \leet\ matrix is present (unfinished receipt). Those SHAs are **pre-pull** values that will always differ from the current checkout — so a single interrupted update permanently poisons the trigger, re-firing a catchup restart on every subsequent \hermes update\ run even though the fleet is already on the right code.

## Fix

Add an early-exit guard in \_receipt_reports_stale_runtime\:

> When \
eceipt.post_update.sha\ (the code version actually pulled by the interrupted run) matches the current checkout SHA, the code **is current** → return \False\ immediately.

The guard is absent when \post_update\ is missing (older receipt format or a run that died before the pull step), preserving the original behaviour for genuinely unfinished updates.

Fixes #98022

### DIFF
--- a/hermes_cli/update_cmd_fleet.py
+++ b/hermes_cli/update_cmd_fleet.py
@@ -122,6 +122,20 @@ def _receipt_reports_stale_runtime(expected_sha: str | None = None) -> bool:
 
     if not _receipt_looks_unfinished(receipt):
         return False
+
+    # Guard (#98022): when the receipt is from an interrupted run but the
+    # code that was *pulled* in that run (receipt["post_update"]["sha"])
+    # already matches the current checkout, the fleet is already on the
+    # right code and no restart is needed. Without this guard an interrupted
+    # receipt triggers a catchup restart on every subsequent `hermes update`
+    # call because `plan.runtimes[].code_sha` always contains the *pre-pull*
+    # SHAs which will never equal the post-pull checkout.
+    post_update = receipt.get("post_update")
+    if isinstance(post_update, dict):
+        pulled_sha = post_update.get("sha") or post_update.get("code_sha")
+        if pulled_sha and str(pulled_sha) == str(expected_sha):
+            return False
+
     plan = receipt.get("plan")
     if not isinstance(plan, dict):
         return False
--- a/tests/hermes_cli/test_update_fleet_stale_receipt_98022.py
+++ b/tests/hermes_cli/test_update_fleet_stale_receipt_98022.py
@@ -1,0 +1,50 @@
+"""Regression tests for #98022 — interrupted receipt must not re-fire restarts.
+
+`plan.runtimes[].code_sha` is captured *before* the pull, so an interrupted
+receipt's plan always looks stale. When the interrupted run already pulled
+the code matching the current checkout (receipt["post_update"]["sha"]),
+the fleet is current and no catchup restart is owed.
+"""
+import hermes_cli.update_cmd_fleet as fleet
+
+
+def _receipt(**over):
+    base = {
+        "stop_reason": "interrupted",
+        "plan": {"runtimes": [{"code_sha": "pre-pull-sha"}]},
+    }
+    base.update(over)
+    return base
+
+
+def test_matching_post_update_sha_reports_current(monkeypatch):
+    monkeypatch.setattr(
+        "hermes_cli.update_receipt.read_latest_receipt",
+        lambda: _receipt(post_update={"sha": "abc123"}),
+    )
+    monkeypatch.setattr(
+        "hermes_cli.update_cmd_fleet._current_checkout_sha", lambda: "abc123"
+    )
+    assert fleet._receipt_reports_stale_runtime() is False
+
+
+def test_mismatched_post_update_sha_still_reports_stale(monkeypatch):
+    monkeypatch.setattr(
+        "hermes_cli.update_receipt.read_latest_receipt",
+        lambda: _receipt(post_update={"sha": "newer-sha"}),
+    )
+    monkeypatch.setattr(
+        "hermes_cli.update_cmd_fleet._current_checkout_sha", lambda: "abc123"
+    )
+    assert fleet._receipt_reports_stale_runtime() is True
+
+
+def test_missing_post_update_preserves_old_behavior(monkeypatch):
+    monkeypatch.setattr(
+        "hermes_cli.update_receipt.read_latest_receipt",
+        lambda: _receipt(),
+    )
+    monkeypatch.setattr(
+        "hermes_cli.update_cmd_fleet._current_checkout_sha", lambda: "abc123"
+    )
+    assert fleet._receipt_reports_stale_runtime() is True


Enough1122:

### PR #98526 — fix(update): skip catchup fleet restart when interrupted receipt pulled SHA already matches checkout

**Summary:** Adds guard in `_receipt_reports_stale_runtime` (#98022): if `receipt["post_update"]["sha"|"code_sha"]` already equals `expected_sha` (current checkout), return `False` — suppresses spurious catchup restart that otherwise fires every run because `plan.runtimes[].code_sha` holds pre-pull SHAs.

- Checks `isinstance(post_update, dict)` and string equality — handles missing/None safely.
- Only applies when receipt looks unfinished (`_receipt_looks_unfinished`) — finished receipts already returned False earlier.
- Single logic, no side effects; `expected_sha` passed from caller is checkout SHA.

**Verdict:** Correct early-exit for poisoned interrupted receipt. No issues.

zengzheqing:
Read the diff: this skips the catch-up when the interrupted receipt's pulled SHA already matches checkout — a reasonable short-circuit, but it only *avoids* the redundant restart; it never discharges the receipt obligation, so the startup warning (`_pending_fleet_restart_needed` → `_receipt_reports_stale_runtime`) keeps firing on every CLI invocation. #103414 stamps `fleet_restart_resolved_sha` onto the owing receipt, which actually clears the warning. The two compose fine (yours prevents the restart, ours clears the debt), but for #98022's reported symptom the receipt discharge is the necessary half.

salch-cred:
Thanks for the review. You're right that the guard only *prevents* the restart but doesn't discharge the receipt — the warning (_pending_fleet_restart_needed → _receipt_reports_stale_runtime) still fires on every CLI invocation.

I see #103414 stamps leet_restart_resolved_sha on the receipt, which discharges it. The two PRs compose: this one prevents the spurious restart, #103414 actually clears the debt by stamping the receipt.

I'll keep this open and add the discharge logic here so #98022's warning is fully resolved in a single PR. (If #103414 lands first, I'll rebase and drop the discharge.)

PR #98528, opened 2026-08-30
fix(relay): call pop_relay_scope directly instead of via run_in_session in _complete_logical

### LINKED ISSUE #97981: relay_llm._complete_logical: run_in_session wrapper breaks logical-parent handles under concurrent turns
While merging a downstream fork sync (base 74ad422d50), `_complete_logical` in `agent/relay_llm.py` wrapped the logical-close in `lease.host.run_in_session(...)`. Under concurrent overlapping turns this invalidated logical-parent handles mid-flight — downstream saw `logical_b is None` assertions and `test_relay_shared_metrics_runtime` failed with handle leakage.

Removing the `run_in_session` wrapper (calling the close directly in the logical context) fixed it: each turn's logical scope is pushed in a dedicated `contextvars.Context` (see `begin_turn`/`_run_context_bounded`), so the logical close must run in *that* context, not re-enter through the session runner which re-binds to a different active turn.

Suggested hardening: either document that `run_in_session` must not wrap logical-scope operations, or make the logical handle lookup immune to concurrent turn re-binding.

salch-cred:
## Problem (#97981)

\_complete_logical\ in \gent/relay_llm.py\ routes \pop_relay_scope\ through \lease.host.run_in_session\ (or \operation_lease.run_in_session\). \
un_in_session\ re-enters the session runner which re-binds the active-turn \ContextVar\ to whatever turn is current **at call time**. Under concurrent overlapping turns that can be a *different* turn — invalidating logical-parent handles mid-flight and leaking them in \logical_llm_calls\.

Downstream symptom: \logical_b is None\ assertions and \	est_relay_shared_metrics_runtime\ failing with handle leakage.

## Fix

Call \pop_relay_scope\ directly. It is a plain function that only operates on the \
elay\ and \handle\ objects passed to it — it does not require session-runner context. Removing the wrapper ensures the close always executes in the logical context that opened it.

Fixes #97981

### DIFF
--- a/agent/relay_llm.py
+++ b/agent/relay_llm.py
@@ -987,12 +987,17 @@ def _complete_logical(
                 output.update({"model": model_name, "provider": provider_name})
                 if response_model_name is not None:
                     output["response_model"] = response_model_name
-            callback = lease.host.run_in_session
-            if operation_lease is not None:
-                callback = operation_lease.run_in_session
-            callback(
-                lease.session,
-                relay_runtime.pop_relay_scope,
+            # Call pop_relay_scope directly rather than routing through
+            # run_in_session.  run_in_session re-enters the session runner
+            # which re-binds the active-turn ContextVar to whatever turn is
+            # current at call time — under concurrent overlapping turns that
+            # can be a *different* turn, invalidating the logical-parent
+            # handle mid-flight and leaking the handle in logical_llm_calls
+            # (#97981).  pop_relay_scope is a plain function that operates
+            # only on the relay/handle objects passed to it; it requires no
+            # session-runner context, so routing it through run_in_session
+            # was never necessary.
+            relay_runtime.pop_relay_scope(
                 lease.host.relay,
                 handle,
                 output=output,


Enough1122:

### PR #98528 — fix(relay): call pop_relay_scope directly instead of via run_in_session in _complete_logical

**Summary:** `_complete_logical` now calls `relay_runtime.pop_relay_scope(lease.host.relay, handle, ...)` directly instead of `lease.host.run_in_session(..., pop_relay_scope, ...)`. `run_in_session` re-binds the active-turn `ContextVar`, so concurrent turns could invalidate `logical_parent` handle mid-flight and leak `logical_llm_calls` (#97981). Direct call is safe since `pop_relay_scope` only operates on relay/handle objects.

- Removes `callback = lease.host.run_in_session / operation_lease.run_in_session` branching — simplifies.
- Comment explains why session-runner context is unnecessary — good for future readers.
- Small diff, no new paths; existing relay tests should still cover handle lifecycle.

**Verdict:** Correct concurrency fix. No issues.

PR #98530, opened 2026-08-30
fix(a2a): set HERMES_CRON_SESSION in subprocess env so approvals respect cron_mode instead of waiting for a phantom terminal

### LINKED ISSUE #98488: [Bug]: A2A-served turns fall through to the interactive CLI approval panel and block for the full approvals.timeout
### Bug Description

When an inbound A2A task trips the dangerous-command gate, the approval prompt
is rendered to a terminal that does not exist instead of being escalated. The
turn stalls for the entire `approvals.timeout` (default 300s) and then returns
`BLOCKED: Command timed out without user response`.

There is no human at a TTY on an s6-supervised gateway, so the prompt can never
be answered. By the time the gate gives up, the A2A caller has already
disconnected, and the peer's completed answer is written to a closed socket:

    BrokenPipeError: [Errno 32] Broken pipe
      File "/opt/hermes/plugins/platforms/a2a/adapter.py", line 306, in do_POST
      File "/opt/hermes/plugins/platforms/a2a/adapter.py", line 195, in _json
        self.wfile.write(body)

Any A2A delegation touching a gated command is guaranteed to fail, and burns a
full timeout plus a complete model turn doing so. Delegating long-running or
specialized work to a dedicated profile — the pattern A2A exists to support —
makes this the common case rather than an edge case.

### Steps to Reproduce

1. Run the official image with two profiles (A and B), each with A2A inbound
   enabled — `gateway.platforms.a2a.enabled: true`, distinct ports.
2. Register B as a peer in A's `a2a_agents`.
3. Leave `approvals` unset on B so the 300s default applies.
4. From A, call B with a task whose tool use trips the dangerous-command
   classifier. In our case B chose a burst cleanup of its own scratch files,
   which fired the `mass_file_deletion` rule:
       rm -f /opt/data/tmp_1.py /opt/data/tmp_2.py /opt/data/tmp_3.py...
5. Observe B's turn hang ~300s at the gate, then return BLOCKED.

### Expected Behavior

An A2A-served turn has no interactive responder, so the gate should resolve
without waiting on one — either by escalating through the gateway approval
path, or by applying a configured policy the way cron sessions already do via
`approvals.cron_mode`, or by failing fast when the selected approval surface
has no reachable responder.

Whatever the resolution, it should happen promptly and the peer should receive
a meaningful reply rather than a dangling socket.

### Actual Behavior

The gate takes the interactive CLI branch and blocks for the full timeout.
Three deadlines then fire in the wrong order and the completed work is
discarded. Measured in a test environment 2026-08-30 (times CEST):

  09:31:34  A -> B   inbound a2a task accepted
  09:32:02  B        terminal tool -> dangerous-command gate -> CLI panel, no TTY
  09:33:34  A        a2a_call timeout (120s) -> "call to 'peer-b' failed — timed out"
  09:36:34  B        _await_reply hits A2A_REPLY_TIMEOUT (300s)
                     -> persists "[agent did not reply in time]"
  09:37:07  B        Tool terminal returned error (305.24s):
                     BLOCKED: Command timed out without user response.
  09:38:05  B        turn completes — 390.8s, 16 API calls, 1020-char answer
                     -> BrokenPipeError writing to the closed caller socket

The 305s gate block is the dominant cost of the 390.8s turn. The peer's
finished answer was unrecoverable — `response_store.db` was empty and only the
`[agent did not reply in time]` placeholder reached `a2a_conversations/`.

Reproduced twice in parallel that morning (305.24s and 306.67s).

### Affected Component

Tools (terminal, file ops, web, code execution, etc.), Other

(The gate itself is `tools/approval.py`; the A2A platform adapter has no
dedicated option — the Gateway entry enumerates Telegram/Discord/Slack/WhatsApp
only.)

N/A (CLI only)

(Literally accurate here: an inbound A2A task is forwarded as a CLI session —
see Root Cause Analysis. A2A is not an option in this dropdown.)

### Debug Report

Not collected via `hermes debug share` for this report — full logs and
tracebacks are included inline below under Additional Logs / Traceback.

### Operating System

Debian GNU/Linux 13 (trixie) — official hermes-agent container; host Debian 12 (bookworm), Linux 6.12.30+

### Python Version

3.13.5

### Hermes Version

Hermes Agent v0.20.6 (2026.8.27) · upstream 2a598aad · install method: docker

Receiving profile, both concurrent tasks:

```
2026-08-30 09:37:07,933 WARNING [20260830_093134_f0a1b5ed] agent.tool_executor: (305.24s): {"output": "", "exit_code": -1, "error": "BLOCKED: Command timed out without user response. The user has NOT consented to this action. Do NOT retry this command..."}
2026-08-30 09:40:06,721 WARNING [20260830_093342_1993f9ec] agent.tool_executor: (306.67s): {"output": "", "exit_code": -1, "error": "BLOCKED: Command timed out without user response..."}
2026-08-30 09:38:05,500 INFO gateway.run: response ready: platform=a2a chat=ctx-epi calls=16 response=1020 chars
```

Calling profile:

```
2026-08-30 09:33:34,803 WARNING agent.tool_executor: Tool a2a_call returned error: call to 'peer-b' failed — timed out.
2026-08-30 09:35:42,395 WARNING agent.tool_executor: Tool a2a_call returned error (120.11s): Error: call to 'peer-b' failed — timed out.
```

Server side, after the caller had gone:

```
Exception occurred during processing of request from ('127.0.0.1', 55384)
Traceback (most recent call last):
  File "/opt/hermes/plugins/platforms/a2a/adapter.py", line 306, in do_POST
    self._json(200, adapter._rpc_message_send(req_id, params, identity, agent=agent))
  File "/opt/hermes/plugins/platforms/a2a/adapter.py", line 195, in _json
    self.wfile.write(body)
  File "/usr/lib/python3.13/socketserver.py", line 845, in write
    self._sock.sendall(b)
BrokenPipeError: [Errno 32] Broken pipe
```

Trigger that fired the gate:

```
{"rule_ids":["mass_file_deletion"],"severity":"CRITICAL","command_redacted":"rm -f /opt/data/tmp_1.py /opt/data/tmp_2.py /opt/data/tmp_3.py...","timestamp":"2026-08-30T07:37:58Z"}
```

An inbound A2A task is forwarded as a CLI session — `adapter.py:867`:

```python
cmd = ["hermes", "chat", "-q", framed_text, "-Q", "--source", "a2a"]
```

so `is_cli=True` for an A2A-served turn. That makes
`_should_fall_through_to_cli_approval()` (`tools/approval.py:329-343`) return True:

```python
return bool(is_cli and approval_callback is not None and notify_cb is None)
```

Control therefore reaches `prompt_dangerous_approval()` (`approval.py:5205`) —
the interactive panel — instead of `submit_pending()` / gateway escalation, and
blocks until `_get_approval_timeout()` expires (default 300s, `approval.py:3432`).

That fall-through is deliberate: its docstring explains it exists to stop
`HERMES_EXEC_ASK` leaking into an interactive CLI and yielding a silent
`pending_approval` with no Approve/Deny UI (cf. #87183). The reasoning holds for
a real terminal, but an A2A-served turn satisfies the same predicate while
having no terminal at all.

Relatedly, `_is_gateway_approval_context()` (`approval.py:293`) exempts cron
sessions via `approvals.cron_mode`, but there is no A2A equivalent — the same
structural gap #87509 reports for `/v1/runs`.

Related issues (none is this branch):
- #87183 (closed) — the CLI-panel fix that introduced this fall-through
- #87488 — headless no-notify_cb path queues into `_pending` with no timer; the
  `is_cli=False` sibling of this branch
- #87509 — `/v1/runs` has no cron_mode-equivalent exemption and stalls
- #84395 / #91980 — gate waits the full timeout when the prompt reaches no one
- #56433 — A2A: wait for final gateway replies before resolving inbound RPCs
- #91687 — A2A long jobs should return a task id

### Proposed Fix (optional)

Any one of, in increasing scope:

1. Exclude A2A-served sessions from the CLI fall-through — gate
   `_should_fall_through_to_cli_approval()` on session source so `--source a2a`
   takes the escalation path rather than the panel. Minimal correct change.

2. Add an A2A exemption to `_is_gateway_approval_context()` mirroring
   `approvals.cron_mode` (e.g. `approvals.a2a_mode: deny|approve|escalate`).
   Gives operators the policy control cron already has, and closes #87509's gap
   in the same place.

3. Fail fast: if the selected approval surface has no reachable responder,
   return immediately instead of waiting out the full timeout. Shared fix with
   #87488 and #84395.

Separately, and arguably a docs matter: `a2a_agents.<peer>.timeout` defaults to
120s (`plugins/platforms/a2a/tools.py:37`) while the receiver's
`A2A_REPLY_TIMEOUT` defaults to 300s (`adapter.py:72`). The sender therefore
always abandons the call before the receiver's own deadline can produce
"[agent did not reply in time]", so that message is written to a socket nobody
is reading. Inverting the default relationship would let the callee's failure
reply actually reach the caller.

salch-cred:
## Problem (#98488)

When an inbound A2A task trips the dangerous-command approval gate, the prompt is rendered to a TTY that does not exist. The turn stalls for the full \pprovals.timeout\ (300 s default), then returns \BLOCKED: Command timed out without user response\. By then the A2A caller has disconnected, causing:

\\\
BrokenPipeError: [Errno 32] Broken pipe
  File '.../plugins/platforms/a2a/adapter.py', line 195, in _json
\\\

## Fix

Set \HERMES_CRON_SESSION=1\ in the A2A subprocess env. \pproval.py\ already respects this to take the \pprovals.cron_mode\ path (default \deny\) rather than waiting for an interactive user. The turn fails immediately with a clear block message; the agent can route around it; no \BrokenPipeError\.

\setdefault\ is used so an existing \HERMES_CRON_SESSION\ in the gateway env is not overwritten. Operators who want auto-approve can set \pprovals.cron_mode: approve\.

Fixes #98488

### DIFF
--- a/plugins/platforms/a2a/adapter.py
+++ b/plugins/platforms/a2a/adapter.py
@@ -875,6 +875,18 @@ class A2AAdapter(BasePlatformAdapter):
             if home:
                 env["HERMES_HOME"] = home
             env["HERMES_A2A_PEER"] = peer
+            # Mark as a non-interactive session so approval.py respects
+            # approvals.cron_mode (default "deny") instead of waiting the
+            # full approvals.timeout (default 300 s) for a human who will
+            # never appear.  Without this, any gated dangerous command
+            # stalls for 300 s, then the A2A caller has disconnected and
+            # the turn reply lands on a closed socket (BrokenPipeError).
+            # Reusing HERMES_CRON_SESSION is intentional: the desired
+            # behaviour is identical — block and let the agent reroute.
+            # Operators who want auto-approve can set
+            # approvals.cron_mode: approve in their config.yaml.
+            # Fixes #98488.
+            env.setdefault("HERMES_CRON_SESSION", "1")
             start = time.time()
             try:
                 proc = subprocess.run(


Enough1122:

### PR #98530 — fix(a2a): set HERMES_CRON_SESSION in subprocess env so approvals respect cron_mode

**Summary:** A2A `_forward_to_profile` now `env.setdefault("HERMES_CRON_SESSION", "1")` so dangerous-command approval gate respects `approvals.cron_mode` (default `deny`) instead of blocking 300 s for a phantom TTY then `BrokenPipeError`.

- Uses `setdefault` — preserves caller-provided value if already set.
- Reuses `HERMES_CRON_SESSION` intentionally; behavior matches cron non-interactive mode (block + reroute). Operators wanting auto-approve can set `cron_mode: approve`.
- Single-line env addition, minimal blast radius.

**Verdict:** Correct non-interactive marking. No blocking issues.

PR #65982, opened 2026-07-16
feat(providers): claude-agent-sdk provider — the official Agent SDK as a first-class runtime under subscription OAuth (fail-closed)

fcavalcantirj:
## Summary

`claude-agent-sdk` as a first-class Hermes runtime: the official Claude Agent SDK under **subscription OAuth**, fail-closed against metered billing. The SDK uses Hermes' existing tool guards and approval surfaces; an explicit SDK provider selection is not silently rerouted to a metered provider. Fixes #25267.

**Review snapshot (2026-09-16):** head `6c580ca1094a` — 41 topic commits on `main@bb0c230366`. GitHub reports the branch mergeable; the PR remains labeled `needs-decision`. Authorship is preserved across rebases: 32 Felipe, 5 CryptoKylan, 4 Romain. For an exact-head re-run, please use the SHA above.

## Review map (41 commits, in order)

- **Foundation and integration (10):** provider extra, config defaults and contributor mappings; Claude Code ↔ Hermes tool identity; SDK session transport and event projector; subscription-lane auxiliary client; gateway approval bridge; hybrid MCP and curated Hermes tools; replay-safe runtime and SDK-owned compaction; background-result delivery; CLI/auth/doctor/session wiring; user and runtime docs.
- **Production and review fixes (5):** Romain's MCP deferral, streamed-prose and image-attachment fixes; CryptoKylan's SDK Bash pre-filter and explicit tirith-allow guard.
- **Footprint and structural split (19):** five main-owned facade seams moved into focused siblings, followed by topic-scoped tests and the transport/runtime splits. The reader stays with the turn; the split does not replace the turn's state gates with a dispatch table. [Split evidence and review map](https://github.com/NousResearch/hermes-agent/pull/65982#issuecomment-5578701543).
- **Post-split hardening and docs (7):** CryptoKylan's terminal-result/late-stop and workspace-bound resume work, the accompanying boundary fixes, the whole-turn stop correction, docs, and the session-schema-history event. These retain their original authorship.

## Conditions and maintainer decision

- Canonical `DEFAULT_CONFIG` entries and tests replaced behavioral `HERMES_CLAUDE_SDK_*` env overrides. The provider-routing F1 found in independent testing is fixed, including the `ANTHROPIC_TOKEN` guard.
[...]

PR #56413, opened 2026-07-01
feat(agent): Claude Agent SDK backend (subscription OAuth) + shared Hermes-tool layer

akshaynexus:
> **Tested end-to-end** with the desktop Electron instance: in `hybrid` mode, Claude Agent SDK inference works — a real Claude-subscription turn ran through the SDK-spawned `claude` CLI (SDK session transcript written under `~/.claude/projects/`, Hermes agent log shows the turn routed through `claude_agent_sdk`'s subprocess transport and returned normally). Feedback still very welcome — especially on the auth story, the three-mode split, and whether the shared tool layer is the right shape.

## What & why

Implements **Refs #25267** — a Claude Agent SDK backend so Claude-subscription users can run Hermes against the [June-15 Agent SDK credit](https://support.claude.com/en/articles/15036540-use-the-claude-agent-sdk-with-your-claude-plan) instead of paying for a separate API key. Drives inference/agent turns through Anthropic's official **`claude-agent-sdk` (Python)** — the SDK spawns the `claude` CLI, which authenticates with a Claude Code / Pro-Max OAuth token (or an API key). Coexists with the existing `anthropic` provider; nothing changes for Developer-Platform users.

Selectable via config:

```yaml
model:
  provider: anthropic
  claude_agent_sdk:
    mode: inference        # inference | delegate | hybrid
    permission_mode: bypassPermissions   # delegate/hybrid
    max_turns: 24
    max_budget_usd: 5.0
```

## Three modes (the #25267 "open implementation question" about two agent loops fighting)

| Mode | Loop driver | Tools |
|------|-------------|-------|
| `inference` (default) | **Hermes** | Hermes' own loop/tools/soul/budget/safety. The SDK is just the model call (single turn, no SDK tools). Text-only — the session-based SDK can't route Hermes' tools on this path. |
| `delegate` | **SDK** | SDK's own Read/Edit/Bash/…, governed by Hermes' guardrails via a `PreToolUse` hook (+ `max_budget_usd`, `disallowed_tools`). |
| `hybrid` | **SDK** | Hermes' tools (registry **and** agent-level, via live `invoke_tool`) exposed through an in-process MCP server, with guardrails and `<untrusted_tool_result>` wrapping. |

## How it plugs in (minimal blast radius)

[...]

PR #38588, opened 2026-06-04
feat: Claude subprocess runtime bridge (/claude-runtime)

atiati82:
## Summary

- Adds `/claude-runtime` slash command that mirrors the existing `/codex-runtime` pattern
- Routes Hermes Anthropic/Claude turns through the official `claude` CLI subprocess
- Bills against the user's Claude Pro/Max subscription tokens instead of consuming API credits
- Full error handling: detects auth failures (401), expired tokens, CLI not installed

## New Files

| File | Purpose |
|------|---------|
| `agent/transports/claude_subprocess.py` | `ClaudeSubprocessSession` — spawns `claude -p --output-format json`, parses results, manages session continuity via `--resume` |
| `agent/claude_runtime.py` | Turn driver (mirrors `codex_runtime.py`) — lazy session init, error recovery, cost tracking |
| `hermes_cli/claude_runtime_switch.py` | `/claude-runtime` slash command — toggle, config persistence, auth status check |
| `hermes_cli/claude_models.py` | Model discovery, CLI binary check, model aliases |

## Wiring

- `conversation_loop.py` — dispatch to claude subprocess when `api_mode == "claude_subprocess"`
- `runtime_provider.py` — resolve `claude_subprocess` from `model.claude_runtime` config
- `gateway/run.py` — handler + running-agent guard (blocks mid-turn switch)
- `commands.py` — command registry entry
- `banner.py` — startup indicator when active
- `run_agent.py` — `AIAgent._run_claude_subprocess_turn()` forwarder

## Usage

```
/claude-runtime on       # enable subscription-based Claude
/claude-runtime off      # revert to API key (default)
/claude-runtime          # show current state + auth status
```

Requires: `claude login` for valid OAuth credentials.

## Test plan

- [x] Syntax check all new/modified files
- [x] Unit test: parse_args, get/set_current_runtime, apply flow
- [x] Integration test: ClaudeSubprocessSession with expired auth (detects 401, actionable error)
- [ ] End-to-end test after `claude login` (pending valid subscription)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #31237, opened 2026-05-24
docs: clarify Anthropic OAuth policy risk

jacobaross:
## Summary
- Recommend `ANTHROPIC_API_KEY` as the safest/default Anthropic setup path.
- Mark Claude/claude.ai OAuth, setup-token, and Claude Code credential reuse as experimental/policy-sensitive unless Anthropic explicitly permits the use case.
- Add guidance to check `hermes auth list anthropic` and remove/suppress OAuth credentials before relying on API-key auth.
- Update provider, quickstart, FAQ, env var, credential pool, OAuth-over-SSH, and provider runtime docs for consistency.

## Why
Anthropic's Consumer Terms and Claude Agent SDK guidance make third-party claude.ai login/rate-limit use policy-sensitive unless explicitly approved. The prior docs could read as if Claude subscription OAuth was a normal recommended path. This makes the safer API-key path clear and avoids implying permission from a working token exchange alone.

## Validation
- `npm ci`
- `npm run build` ✅ exited 0
  - Existing/pre-existing Docusaurus broken-link and broken-anchor warnings remain, especially in localized docs.
  - Prebuild skipped populated Skills Hub extraction because local Python lacks `yaml`, but the build completed.
- `git diff --check` ✅
- `npm run lint:diagrams` ⚠️ could not run locally because `ascii-guard` is not installed/available after `npm ci`.

PR #98533, opened 2026-08-30
feat: claude-code-cli provider — Claude subscription via the Claude Code CLI with real tool calls

### LINKED ISSUE #25267: [Feature]: Claude Agent SDK model provider with subscription OAuth (Codex-style)
### Problem or Use Case

I want to run Hermes with Claude as the model backend while staying on my Claude subscription, but the existing `anthropic` provider requires a Developer Platform API key and bills separately — so Claude-subscribed users effectively pay twice (subscription + per-token API).

Starting **June 15, 2026**, Anthropic decouples Agent SDK and `claude -p` usage from interactive subscription limits and introduces a separate monthly Agent SDK credit:

| Plan | Monthly Agent SDK credit |
|---|---|
| Pro | $20 |
| Max 5x | $100 |
| Max 20x | $200 |
| Team (Standard / Premium) | $20 / $100 |
| Enterprise (usage-based / Premium seats) | $20 / $200 |

The credit explicitly covers *"third-party apps that authenticate with your Claude subscription through the Agent SDK."* Source: [Anthropic Help Center](https://support.claude.com/en/articles/15036540-use-the-claude-agent-sdk-with-your-claude-plan).

Today there's no way to route Hermes into that credit — even Pro / Max subscribers must fund an API key on top of their plan. A Claude-subscription-OAuth provider would close the gap, the same way the existing Codex OAuth credential pool already does for OpenAI subscribers.

### Proposed Solution

A new model-provider plugin that wraps the Claude Agent SDK with Claude subscription OAuth, mirroring the existing Codex OAuth path. Coexists with the current `anthropic` (API-key) provider rather than replacing it, so users on the Developer Platform are unaffected.

**Plugin layout**
- New plugin at `plugins/model-providers/claude-agent-sdk/`, wrapping `claude-agent-sdk` (Python) or `@anthropic-ai/claude-agent-sdk` (TypeScript).
- Either a new `ClaudeAgentSdkTransport` under `agent/transports/`, or extend the existing `AnthropicTransport` to swap auth header / endpoint by credential type — whichever fits the `ProviderTransport` ABC more cleanly.

**Auth integration**
- Detect Claude Code OAuth credentials at `~/.claude/`, paralleling the existing `~/.codex/auth.json` import path.
- Reuse the Codex credential-pool plumbing for refresh, rotation, and stale-token handling.

**Config / UX**
- New provider option (e.g. `claude_subscription`) selectable in `hermes model`, symmetric to the Codex subscription option.
- New env var, e.g. `HERMES_CLAUDE_SUBSCRIPTION_AUTH=1`, or auto-detect when `~/.claude/` credentials exist and no Anthropic API key is set.
- Surface a docs / UX warning when a subscription provider is selected for shared multi-user deployments — per Anthropic, the credit is *"sized for individual experimentation"*; production multi-tenant automation should still use API keys.

**Open implementation questions**
- The Agent SDK ships its own agent loop. Since Hermes is itself an agent, the cleanest integration is likely the SDK's lower-level / non-interactive surface rather than letting both loops fight for control of tool use and context. Worth scoping during design.
- Anthropic's formal terms for subscription auth in third-party Agent SDK apps will be posted when the claim flow opens — worth re-checking against Hermes' usage pattern before merging.

**References**
- Help Center: https://support.claude.com/en/articles/15036540-use-the-claude-agent-sdk-with-your-claude-plan
- Agent SDK overview: https://code.claude.com/docs/en/agent-sdk/overview
- Claude Agent SDK (Python): https://github.com/anthropics/claude-agent-sdk-python

### Alternatives Considered

- **Keep the status quo (`anthropic` provider with API key).** Works, but Claude-subscribed users pay twice. The new June 15 credit policy makes this less defensible: there's now a published, supported path for third-party apps to authenticate through the subscription.
- **Route Claude via OpenRouter.** Adds ~5.5% platform fee on top, still pay-per-token, and doesn't touch the new subscription credit. Useful for multi-provider flexibility, not for cost.
- **Switch to Codex / OpenAI subscription via the existing Codex OAuth pool.** Already supported and works today, but forces users away from Claude as the backend, which isn't a trade-off Claude-subscription users would want to make.
- **Community subscription-auth shims (`claude_max`, `claude-code-router`, etc.).** Technically bolt-on-able, but lives in ToS gray-area territory and is likely to break as Anthropic tightens auth flows around the formal credit program. A first-class provider plugin is cleaner and more durable.

### Feature Type

Other

### Scope

None

### Contribution

### Debug Report (optional)

```shell

```

Izzy-Gottz:
## What does this PR do?

Adds a `claude_code` api_mode and a `claude-code-cli` provider so Hermes can use a **Claude subscription with real tool calls**, by driving the official Claude Code CLI (`claude -p --input-format stream-json --output-format stream-json`) as a long-lived, Hermes-owned subprocess — the same shape as the existing `codex_app_server` runtime. Hermes's tools are exposed to the child over stdio MCP (`hermes_tools_mcp_server`), so Claude calls `web_search`, `browser_*`, etc. natively; Claude Code's own tools (Bash, Read, Edit…) keep working.

Why this approach: subscription OAuth cannot be used directly against the API (Anthropic's consumer terms), and the community proxy route strips tool calls and starts a cold process per request. Spawning the official CLI is the path Anthropic's own docs describe; only the CLI ever touches the credential.

The child is **isolated from the user's personal Claude Code state**: its own `CLAUDE_CONFIG_DIR` under `$HERMES_HOME/claude-code`, its own cwd, `--setting-sources ""` + `--settings <Hermes-written deny list>`, `--strict-mcp-config`, auto-memory off. It authenticates with a long-lived setup-token (`claude setup-token` → `CLAUDE_CODE_OAUTH_TOKEN`), never the interactive keychain login — which also avoids refresh-token rotation races with the user's own `claude` sessions.

Permission modes map from `tools.terminal.security_mode`: `auto` → `acceptEdits` + deny list; `approval-required` → `default` with `--permission-prompt-tool stdio` routed through Hermes's approval callback (`can_use_tool` control requests); `unrestricted`/`yolo` → `bypassPermissions`. Unknown values fail closed.

On the `api_server` gateway path (fresh `AIAgent` per request) a module-level registry keeps one warm child per Hermes session (per-session turn lock, idle/LRU eviction, atexit cleanup), and the request's ephemeral system prompt is baked into the child's `--append-system-prompt-file` (respawn + `--resume` when it changes).

## Related Issue

Fixes #25267

## Type of Change

## Changes Made

- `agent/transports/claude_code_session.py` — `ClaudeCodeSession`: spawn, warm-up-before-init, stream-json projection (text, tool_use/result, usage), drain-before-finish, process-identity checks, interrupt, `--resume` with session-id rotation only on the CLI's own rejection, permission-prompt bridge, isolation flags/env, temp files 0600.
- `agent/claude_code_runtime.py` — `run_claude_code_turn` (same return shape as `run_codex_app_server_turn`), UI event bridge, usage accounting, warm-session registry, `combined_system_prompt`, `codex_app_server` site audit in the docstring.
- `plugins/model-providers/claude-code/` — `claude-code-cli` ProviderProfile (aliases `claude-subscription`, `claude_code_cli`). `claude-code` remains an alias for the `anthropic` API-key provider (unchanged).
- Wiring: `agent/conversation_loop.py`, `run_agent.py`, `agent/agent_init.py`, `agent/turn_context.py`, `agent/conversation_compression.py`, `agent/background_review.py` (review skipped on this runtime — its tools can't run over stateless MCP), `hermes_cli/{auth,models,providers,runtime_provider}.py`.
- Tests: `tests/agent/transports/{fake_claude_cli.py,test_claude_code_session.py}`, `tests/agent/test_claude_code_runtime.py`, `tests/hermes_cli/test_claude_code_cli_provider.py` — fake CLI speaking stream-json; no network.
- Docs: `website/docs/integrations/providers.md` (en) + zh-Hans section; `cli-config.yaml.example` `claude_code:` block.

## How to Test

1. Install Claude Code (`claude --version` ≥ 2.1), sign in, run `claude setup-token` and `export CLAUDE_CODE_OAUTH_TOKEN=...`.
2. `hermes chat -Q --oneshot --provider claude-code-cli --model sonnet -q "Reply with exactly: OK"` → `OK`.
3. Tool call through MCP: `... -q "Use the web_search tool to find the current Swift version, reply with just the number"` — verbose output shows `tool_use: web_search` and the answer.
4. Isolation: after a turn, `~/.claude/projects/*/memory` is untouched; the child's transcript is under `$HERMES_HOME/claude-code/projects/`.
5. Deny list: in `auto` mode ask it to `git push` in a scratch repo → `Permission to use Bash … has been denied`, remote unchanged.
6. `pytest tests/agent/transports/test_claude_code_session.py tests/agent/test_claude_code_runtime.py tests/hermes_cli/test_claude_code_cli_provider.py -q` → 56 passed.

### Code

- [x] I've run `pytest tests/ -q` — full run via `scripts/run_tests.sh`: no regressions versus the base commit (127 failures, all pre-existing on upstream main in an offline environment; details in the comment below)
- [x] I've tested on my platform: macOS 15 (Darwin 25.5), Claude Code 2.1.251, Python 3.11

- [x] I've updated relevant documentation
- [x] I've updated `cli-config.yaml.example`
- [x] `CONTRIBUTING.md`/`AGENTS.md` — N/A (no workflow change)
- [x] Cross-platform: the runtime requires the `claude` CLI; macOS/Linux tested paths only. Windows untested — the CLI exists there but the process-group and pipe handling have not been exercised.

## Notes for reviewers

- `approval-required` mode relies on `--permission-prompt-tool stdio`, which is absent from `claude --help` but present in the 2.1.251 binary and verified live; if a future CLI drops it, the mode falls back to read-only tools with a loud status line.
- Known, inherent to the CLI: `CLAUDE_CODE_OAUTH_TOKEN` is visible to shell commands the model runs (Hermes's terminal tool deliberately doesn't strip it either). Documented in the runtime; containment is the deny list + a revocable setup-token.
- Background memory/skill review is skipped on this runtime (its tools are `_AGENT_LOOP_TOOLS` and can't be exposed over stateless MCP); a stateless memory tool over MCP is a possible follow-up.

### DIFF
--- a/agent/agent_init.py
+++ b/agent/agent_init.py
@@ -522,10 +522,12 @@ def _refuse_checkpoint_required_on_codex_app_server(
     compress_context() guard alone cannot cover native turns that bypass
     Hermes compression entirely.
     """
-    if checkpoint_required and api_mode == "codex_app_server":
+    # claude_code has the same shape: the `claude` CLI compacts its own
+    # context and Hermes never sees a pre-compaction boundary (#25267).
+    if checkpoint_required and api_mode in {"codex_app_server", "claude_code"}:
         raise RuntimeError(
             "BLOCKED_MISSING_PREREQUISITE: compression.checkpoint_required "
-            "is incompatible with the codex_app_server API mode: the codex "
+            f"is incompatible with the {api_mode} API mode: the codex "
             "agent compacts its own thread without a truthful pre-compaction "
             "transcript boundary, so a required pre-compress checkpoint "
             "cannot be guaranteed. Disable compression.checkpoint_required "
@@ -715,8 +717,10 @@ def init_agent(
     agent._credential_pool = credential_pool
     agent.acp_command = acp_command or command
     agent.acp_args = list(acp_args or args or [])
-    if api_mode in {"chat_completions", "codex_responses", "anthropic_messages", "bedrock_converse", "codex_app_server"}:
+    if api_mode in {"chat_completions", "codex_responses", "anthropic_messages", "bedrock_converse", "codex_app_server", "claude_code"}:
         agent.api_mode = api_mode
+    elif agent.provider == "claude-code-cli":
+        agent.api_mode = "claude_code"
     elif agent.provider == "openai-codex":
         agent.api_mode = "codex_responses"
     elif agent.provider in {"xai", "xai-oauth"}:
@@ -1283,6 +1287,16 @@ def init_agent(
         if not agent.quiet_mode:
             _gr_label = " + Guardrails" if agent._bedrock_guardrail_config else ""
             print(f"🤖 AI Agent initialized with model: {agent.model} (AWS Bedrock, {agent._bedrock_region}{_gr_label})")
+    elif agent.api_mode == "claude_code":
+        # Claude Code subscription runtime — the `claude` CLI owns the network
+        # and authenticates with its setup-token. No HTTP client, no API key.
+        # See agent/claude_code_runtime.py (#25267).
+        agent.client = None
+        agent._client_kwargs = {}
+        agent.api_key = api_key or "claude-code-subscription"
+        agent.base_url = base_url or "claude-code://local"
+        if not agent.quiet_mode:
+            print(f"🤖 AI Agent initialized with model: {agent.model} (Claude Code subscription)")
     else:
         client_kwargs = {}
         if api_key and base_url:
--- a/agent/background_review.py
+++ b/agent/background_review.py
@@ -1413,6 +1413,21 @@ def _run_review_in_thread(
     if review_run is not None and review_run.cancel_requested.is_set():
         finish_background_review_run(agent, review_run)
         return
+    # claude_code runtime: the fork would spawn a second `claude` process that
+    # cannot reach the `memory` / `skill_manage` tools (they are
+    # _AGENT_LOOP_TOOLS, not exposable over MCP), so the review could never
+    # save anything. Refuse here — every caller (automatic turn-boundary
+    # review, /refine, the CLI review command) converges on this thread
+    # target (#25267).
+    if getattr(agent, "api_mode", None) == "claude_code":
+        logger.info(
+            "background review skipped: api_mode=claude_code cannot write "
+            "memory/skills from the claude subprocess (session=%s)",
+            getattr(agent, "session_id", None),
+        )
+        if review_run is not None:
+            finish_background_review_run(agent, review_run)
+        return
 
     # Local import to avoid a hard circular dep at module load.
     from run_agent import AIAgent
--- a/agent/claude_code_runtime.py
+++ b/agent/claude_code_runtime.py
@@ -1,0 +1,869 @@
+"""Claude Code runtime — hand a turn to a ``claude -p`` subprocess (#25267).
+
+Mirror of :mod:`agent.codex_runtime.run_codex_app_server_turn` for the
+``claude_code`` api_mode: the official Claude CLI owns the model loop and its
+native tools, Hermes' tools reach it over stdio MCP, and the CLI's stream-json
+events are projected back into Hermes' ``messages`` list so memory review,
+session persistence, and every gateway keep working unchanged.
+
+Sites that string-compare ``"codex_app_server"`` and what this runtime does
+about each (grep ``codex_app_server`` to audit):
+
+  MIRRORED
+    agent/conversation_loop.py   dispatch           -> ``elif api_mode == "claude_code"``
+    run_agent.py                 _run_*_turn         -> ``_run_claude_code_turn`` forwarder
+    run_agent.py                 interrupt()         -> forwards to ``_claude_code_session.request_interrupt``
+    run_agent.py                 close()             -> closes ``_claude_code_session``
+    agent/agent_init.py          valid api_mode set  -> ``"claude_code"`` added
+    agent/agent_init.py          checkpoint refusal  -> Claude Code compacts its own context, same as codex
+    agent/agent_init.py          client construction -> no OpenAI client; ``agent.client = None``
+    agent/turn_context.py        preflight compress  -> skipped (native compaction; == ``codex_app_server_auto: native``)
+    agent/turn_context.py        api-bytes stamp     -> skipped (the CLI never sees ``api_messages``)
+    agent/conversation_compression.py compress_context -> no-op boundary (native compaction only)
+    agent/background_review.py   _run_review_in_thread -> refused on claude_code for EVERY caller
+                                                        (automatic, /refine, CLI command): the fork
+                                                        would start a second `claude` that cannot
+                                                        reach the `memory` / `skill_manage` tools
+                                                        (they are _AGENT_LOOP_TOOLS, not MCP-exposable)
+    tui_gateway/server.py        image_routing       -> forces text mode for codex_app_server; claude
+                                                        images are not forwarded either (see
+                                                        _coerce_input_text) so the same downgrade applies
+    hermes_cli/runtime_provider.py _VALID_API_MODES  -> ``"claude_code"`` added; ``claude-code`` provider resolves
+                                                        with no base_url / api_key
+
+  NOT MIRRORED (deliberately)
+    run_agent.py                 steer()             -> codex has ``turn/steer``; stream-json has no
+                                                        equivalent. Falls through to the default
+                                                        next-turn queue.
+    agent/conversation_compression.py _compress_context_via_codex_app_server
+                                                     -> ``compression.codex_app_server_auto: hermes``
+                                                        has no analogue; Claude Code exposes no compact RPC.
+    agent/native_compaction.py                       -> codex-only accounting for the above.
+    hermes_cli/codex_runtime_switch.py, commands.py, banner.py, cli.py /codex-runtime,
+    gateway/slash_commands.py, codex_runtime_plugin_migration.py
+                                                     -> codex-specific opt-in UX (``model.openai_runtime``).
+                                                        claude_code is selected by provider, not a switch.
+    hermes_cli/config_defaults.py, gateway/run.py    -> ``compression.codex_app_server_auto`` config key.
+                                                        claude_code behaves as the ``native`` value of
+                                                        that knob, permanently: the CLI auto-compacts
+                                                        its own context, Hermes never summarizes the
+                                                        local mirror, and there is no ``hermes``/``off``
+                                                        analogue because the CLI exposes no compact RPC.
+    agent/transports/hermes_tools_mcp_server.py      -> docstring mentions only; shared as-is.
+
+Known gap shared with the codex runtime: Hermes tools that need the live
+agent loop (``memory``, ``delegate_task``, ``session_search``, ``todo``) are
+not exposed over MCP.
+
+TODO: background memory/skill review is skipped on this runtime (the guard
+lives in ``agent/background_review._run_review_in_thread`` so every caller —
+automatic, ``/refine``, CLI command — converges on it). The follow-up that
+would restore it is a stateless ``memory`` MCP tool backed by a fresh
+``MemoryStore`` loaded from disk per call, with the parent reloading its
+store after each turn.
+
+Isolation contract (HIGH-2 in review): the child is a Hermes-owned Claude
+Code — ``CLAUDE_CONFIG_DIR=$HERMES_HOME/claude-code``, cwd
+``$HERMES_HOME/claude-code/workspace`` (or ``claude_code.cwd``),
+``--setting-sources ""`` + ``--settings $HERMES_HOME/claude-code/settings.json``
+(deny list), ``--strict-mcp-config``, auto-memory off, and its own
+setup-token credential in ``$CLAUDE_CODE_OAUTH_TOKEN``. The user's
+``~/.claude`` (hooks, plugins, memory, transcripts, interactive login) is never
+read or written.
+"""
+
+from __future__ import annotations
+
+import atexit
+import glob
+import logging
+import os
+import threading
+import time
+import uuid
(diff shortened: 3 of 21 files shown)


andrexibiza (commented):
Reviewed exact head `287fd8083e938855fda02b8c0b0c422e4a2caaaf`. The subprocess lifecycle work is careful — especially the warm-up-before-init protocol, process-identity check on reader EOF, callback rebinding for the warm registry, resume rotation only on explicit CLI rejection, and the real-pipe fake-CLI harness. I would not land this head yet, though. There are two architecture blockers, one functional gap, and no hosted acceptance on the reviewed object.

## P1 — the native Claude tool surface bypasses Hermes' safety + secret boundary

The normal `auto` path is materially less constrained than Hermes' existing tool path:

- `resolve_permission("auto")` returns `acceptEdits` with `Bash`, `Read`, `Write`, `Edit`, `MultiEdit`, `NotebookEdit`, `Task`, etc. in `_AUTO_MODE_ALLOWED_TOOLS`; `_build_command()` forwards them in `--allowedTools`. `test_permission_mapping` and `test_mcp_config_flag_and_allowlist` explicitly pin `Bash` as pre-approved.
- `build_child_env()` intentionally starts from `hermes_subprocess_env(inherit_credentials=True)` because the nested Hermes MCP server needs provider credentials, and then adds the long-lived `CLAUDE_CODE_OAUTH_TOKEN`. The source correctly notes that `env` inside Claude's native Bash can read that token.
- Those **native** Bash/Read/Write/Edit calls never traverse Hermes' canonical guards. Current `tools/terminal_tool.py` routes ordinary terminal execution through `tools.approval.check_all_command_guards()` (Tirith + dangerous-command detection + the session-aware approval policy), and Hermes' file path has its own sensitive-path boundary. A Claude-native tool call bypasses both. A short static `permissions.deny` list is not equivalent to those policies.
- `CLAUDE_CONFIG_DIR` and a dedicated cwd isolate Claude Code *state*; they do not sandbox the OS principal. Native `Read` can address absolute paths, native `Write/Edit` run as the Hermes user, and native `Bash` can read inherited environment variables and make network calls. Output redaction helps accidental display; it does not prevent effect/exfiltration.

That means a prompt-injected or simply mistaken model in the default `auto` configuration can cross a boundary Hermes currently owns: it can execute outside Tirith/dangerous-command admission, reach protected filesystem paths outside the dedicated cwd, and observe/exfiltrate the setup token plus any inherited provider credentials. The code comments acknowledge the token visibility, but the current mitigation does not close the authority gap.

**Required before merge:** make Hermes the authority owner for native mutating/executable effects. A few acceptable shapes exist, but the invariant should be unambiguous: normal modes cannot pre-authorize an OS-level tool that holds Hermes credentials while bypassing Hermes' command/file policy. Either route those Claude tool requests through the canonical Hermes guards/scoped execution path, or disable the native mutating/executable surface under normal modes and expose Hermes-controlled equivalents; reserve truly unguarded native execution for the operator's explicit unrestricted/yolo choice. Add adversarial regressions that prove (1) a Bash command caught by Hermes dangerous/Tirith logic cannot execute pre-effect in `auto`, (2) protected file reads/writes cannot bypass Hermes' path policy, (3) native execution cannot observe/exfiltrate `CLAUDE_CODE_OAUTH_TOKEN` or inherited provider credentials in normal modes, and (4) the allowed Hermes MCP tool path still works.

There is useful prior design work here: `akshaynexus`'s open #56413 explicitly routes the SDK-owned tool surface through a Hermes `PreToolUse` guardrail and centralizes shared Hermes-tool exposure. That PR is a different runtime architecture, not something to copy blindly, but its authority-boundary work is directly relevant to this gap and should be preserved/credited if #98533 becomes the convergence object.

## P1 — this is a competing owner for #25267, but the provenance/merge-order graph is currently absent

Live #25267 (reported by `YongboYu`) is still open and labeled `needs-decision`. #98533 says the duplicate search was completed and uses `Fixes #25267`, but the PR does not identify the substantial open work already occupying this exact design space:

- `atiati82`'s #38588 is the direct raw-`claude -p` subprocess predecessor. It is still open and collides with #98533 in `agent/conversation_loop.py`, `hermes_cli/runtime_provider.py`, and `run_agent.py`.
- `akshaynexus`'s #56413 is the official Agent SDK alternative, with an explicit map of competing approaches and the shared-tool/guardrail work above.
- `fcavalcantirj`'s #65982 is another first-class Agent SDK/provider implementation for #25267, with independent live Max-subscription verification and a substantial billing/routing hardening history.
- `jacobaross`'s #31237 is the still-open policy/terms clarification already cross-linked from #25267.

Those are not all duplicates of each other, but they are not unrelated either. #38588 is direct competing mechanism/ownership; #56413 and #65982 are competing architectural realizations with useful complementary work; #31237 is policy lineage. Landing #98533 without an explicit convergence decision would create two raw subprocess owners plus multiple provider owners for the same feature and silently erase prior contributor work from the new canonical path.

**Required before merge:** add an explicit lineage/convergence section that distinguishes predecessor, competing, complementary, and superseded work; preserve the authors above and any additional discovered contributors; decide which PR is the canonical implementation; and state merge/closure order. If #98533 supersedes #38588, say exactly what is subsumed and what is not. If it replaces the SDK approaches, explain the architectural decision and preserve any guardrail/tool-exposure/billing work that remains independently valuable. Because #25267 still carries `needs-decision`, I would also avoid auto-closing it until that repository-level decision is actually resolved.

## P2 — image routing is documented as mirrored, but the live gateway does not mirror it

`agent/claude_code_runtime.py` says the `tui_gateway/server.py` image-routing special case is mirrored because Claude images are not forwarded. The reviewed diff never changes `tui_gateway/server.py`, and current main only forces `_mode = "text"` when `agent.api_mode == "codex_app_server"`. Meanwhile `ClaudeCodeSession._coerce_input_text()` explicitly replaces an `image_url` block with `[image attached — not forwarded to claude]`.

So the stated parity is not present at the routing owner. If the normal image-routing resolver selects a vision/multipart path for a Claude model, this runtime can silently turn an attachment into a placeholder rather than establishing the same explicit text-only downgrade as Codex.

**Required:** either add `claude_code` to the canonical text-only routing decision or actually forward supported image content, and add the gateway-level regression that proves an attachment cannot be silently dropped.

## Structure / acceptance gates

This feature adds more runtime string-compare branches to existing godfiles: `agent/conversation_loop.py` is already past line 2,025 at the new dispatch, and `run_agent.py` is touched around lines 3,376 / 4,716 / 9,119. This is exactly the kind of cross-runtime lifecycle/dispatch ownership that should be extracted behind a bounded runtime interface instead of accreting another branch into >2K owners. Please use the new provider to pull the generic runtime dispatch/interrupt/close seam out rather than growing those files again.

Exact-object acceptance is also absent right now. This head has four commits; the first three have no workflow runs at all, and the head's CI `33309498629`, Docker `33309498122`, and Nix `33309498208` are all `action_required`; the CI run has **zero jobs**. The PR body itself still says the full suite is in progress and Windows is untested. The 56 targeted fake-CLI tests are useful construction evidence, but they are not a hosted landing receipt for a new process/auth/permission protocol. The current head is now `4 ahead / 2 behind` live `main@c49fa88b80753071e7b7bb83e2882232e43e87c0`, so the repaired convergence object needs current-main recomposition and fresh exact-head CI.

Attribution CI has not run either. The four commits use `Izzy-Gottz <srulynj@gmail.com>` and current main has no `contributors/emails/srulynj@gmail.com` mapping, so preserve that identity and add the mapping rather than rewriting authorship when the branch is repaired.

The process/session mechanics here are worth keeping. Close the native-tool authority leak, converge the existing #25267 implementations with credit intact, fix the image-routing owner, extract the runtime seam instead of extending the godfiles, and then prove the rebuilt exact head through hosted CI. At that point this can be a strong Claude-subscription runtime rather than a parallel safety model sitting beside Hermes' own one.

Izzy-Gottz:
**Test run on head `287fd80`** — full suite (`scripts/run_tests.sh`, ~35.5k tests, macOS, no network): 127 failures, all in files untouched by this PR. Re-ran those same 41 files on the base commit `2a598aa` without the patch: 134 failures, a superset of the above. The 3 tests that failed only with the patch (`tests/agent/lsp/test_client_e2e.py` ×2, `tests/agent/test_shell_hooks_tree_kill.py::test_timeout_kills_descendants`) pass in isolation on the PR branch — load-sensitive, not regressions. The 67 targeted transport/runtime/provider tests pass.

@andrexibiza — thank you for the thorough review. The native-tool authority gap (P1) is the right call and I intend to fix it at the root — Hermes owns mutating/executable effects; native `Bash/Write/Edit` off under normal modes, Hermes MCP equivalents on — rather than widen the deny list. I'll follow up on the lineage/convergence section, the image-routing owner, and the dispatch seam, and will re-request review on a rebased head with CI.

simonyos:
Hi — I've been building almost exactly this over the last week without having seen your PR, so reading the summary was a bit surreal. api_mode, the codex_app_server shape, tools over stdio MCP, --strict-mcp-config, and the whole "only the CLI touches the credential" argument. Same conclusions, independently.

Two things from that work that might be useful here:

The read/write flag table in hermes_tools_mcp_server is the spot I'd want a test on. Anything flagged false gets answered with no prompt at all, so one wrong entry there is a silent auto-approval rather than a loud failure.

The approval path was fail-open when I got to it. Everything that wasn't literally "deny" or "timeout" granted, so a callback handing back "" or "no" or None approved the command. Filed that one separately as #109266 because it isn't CLI-specific. While doing the CLI leg I also found two of the three decision points still granted on an unrecognised answer, so guarding just the CLI callback wasn't enough in my tree.

I also grew the tool catalog so a CLI session can actually do what a native one does (browser, kanban, vision, tts, with the read-only ones auto-answered), and wrote the architecture up with a file:line for every claim.

Happy to paste either, or send patches against your branch, if that's useful. Not trying to open a competing PR.

Issue #95976, opened 2026-08-27
Background review fork fails to update skills 100% of the time — read-before-write guard conflicts with skill_view repeat-view dedup


yaozhen:
## Title

**Background review fork fails to update skills 100% of the time — read-before-write guard conflicts with `skill_view` repeat-view dedup**

## Summary

The autonomous background review fork (`agent/background_review.py`) is *supposed* to update skills after each turn. In practice it **never succeeds at updating a skill**: every `patch`/`write_file` against an existing skill is refused by the read-before-write guard, because the fork's `skill_view` call hits the repeat-view dedup first and returns a `content_returned: false` stub that never marks the file as read. Result: the fork spins up an LLM, tries to patch, gets refused, exits — every time, wasting tokens with zero skill updates landing.

## Environment

- Hermes **v0.20.5** (the behavior reproduces on current `main`, e1e72f109c)
- edittable install; macOS; single gateway process

## Evidence (production, 2026-07-24 → 08-26)

From `errors.log`:

```
Refusing background curator patch for skill 'X': the current SKILL.md content
has not been loaded in this review turn. Call skill_view(name) for SKILL.md, …
```

- **408** such refusals over the period (49 on a single day 08-26).
- Skill-completion telemetry (`Background review complete: … result=…`): **41 forks completed, `result=skill` appeared 0 times**; `result=memory` appeared 9 times, `result=none` 32 times. Memory updates work; skill updates never do.

## Root cause

Two mechanisms that each make sense in isolation, but are structurally incompatible in the fork:

1. **Read-before-write guard** — `tools/skill_manager_tool.py::_background_review_read_before_write_guard` rejects `patch`/`write_file` unless `_background_review_has_read()` is true. That flag is set only by `mark_background_review_skill_read()`, which `skill_view` calls **only on the path that actually reads and returns file content** (`tools/skills_tool.py`).

2. **Repeat-view dedup** — `tools/skills_tool.py::_skill_view_tracker` (module-level dict guarded by `threading.Lock`) returns a `content_returned: false` stub when the same task already viewed an unchanged file. On that path `mark_background_review_skill_read` is **never called**.

3. **The fork shares the parent's session** — `agent/background_review.py` sets `review_agent.session_id = agent.session_id` (for prefix-cache parity). The dedup cache is keyed by `task_id` (== `session_id` in the gateway) and lives in the shared module, so the fork's `skill_view` almost always hits the dedup stub for skills the parent already loaded.

[...]

PR #98538, opened 2026-08-30
fix(agent): add absorbed_into to skill_manage schema (#97959)

### LINKED ISSUE #97959: Curator consolidation can never complete: absorbed_into is accepted by the skill_manage handler but omitted from the tool schema
### Summary

The curator's LLM consolidation pass (`agent/curator.py`, `CURATOR_REVIEW_PROMPT`) instructs the review agent to archive absorbed skills with `skill_manage action=delete, absorbed_into=<umbrella>`. The handler accepts `absorbed_into` (`tools/skill_manager_tool.py`, `_delete_skill()`), but the parameter is **deliberately omitted from `SKILL_MANAGE_SCHEMA`** (see the `# NOTE:` block above `"required": ["action", "name"]`). Models that strictly follow the tool schema therefore never emit it, every delete is refused by `_curator_consolidation_delete_guard()`, and the pass ends with **zero consolidations while reporting success**.

### Environment

- Hermes Agent v0.20.5 (2026.8.19), upstream `6defe7eb`, docker install
- Main + auxiliary models via an OpenAI-compatible community router (non-frontier models)
- 76-skill local library, `curator.consolidate` triggered via `hermes curator run --sync --consolidate`

### Reproduction (live run, 2026-08-27)

1. Disabled `tool_loop_guardrails` (warnings + hard stop) to isolate the behavior — the previous run (2026-08-24) had been halted by `same_tool_failure_halt` after 8 refused deletes.
2. Ran `hermes curator run --sync --consolidate`. The review agent ran to completion (81 tool calls, no guardrail halt).
3. Result (from `logs/curator/<run>/run.json`, machine-readable):
   - 7 umbrella skills **created** successfully (`ai-model-economics`, `hermes-operations`, `memory-systems`, `whatsapp-operations`, `research-recovery`, `media-visual-operations`, `playlist-operations`)
   - **19 consecutive `skill_manage(action="delete")` calls, every single one refused** by `_curator_consolidation_delete_guard` with *"Pass absorbed_into=<umbrella> ..."*
   - `consolidated_this_run: 0`, `archived_this_run: 0`, `pruned_this_run: 0`
   - Run classified as success; final summary: *"Completed the first umbrella-building pass and verified the catalog afterward."*

Key detail: the refusal error message **explicitly names `absorbed_into`**, and the agent still issued a second full wave of deletes (calls 45–80 in `tool_calls`) without the parameter. The schema gap, not prompt wording, is what binds: the JSON schema the model sees has no `absorbed_into` property at all.

Additional failure mode: on gateways that deduplicate function-calling tool schemas by parameter name (several OpenAI-compatible community routers merge identically-named properties into one), the model cannot pass `absorbed_into` on a delete call at all — `action` and `name` already appear as top-level keys, so a second occurrence of an existing key name pattern gets collapsed. Even a schema fix may not be reachable through such gateways unless the parameter name is unique per call shape.

### Root cause

`tools/skill_manager_tool.py`:

```python
# NOTE: the handler also accepts `absorbed_into` on delete — the
# curator's consolidation pass declares merge-vs-prune intent with
# it. Deliberately NOT advertised in this schema: only curator
# sessions need it, the curator's own prompt documents it, and the
# curator-context delete guard's error re-teaches it on omission
# (_curator_consolidation_delete_guard). Keeping it out saves
# ~100 tokens on every call of every other session.
```

The token-saving assumption ("the curator's prompt documents it") is empirically false for non-frontier models: the schema is the stronger constraint, the prompt cannot reliably override it, and the guard's error message does not recover the call either.

### Why this is a silent failure

- The run exit code is 0 and `llm_error` is null.
- The LLM summary reads as success ("completed... verified the catalog").
- Nothing in `curator status` distinguishes "consolidation complete" from "created umbrellas but archived nothing."
- Meanwhile the library is left in the worst state for discoverability: new umbrella skills exist alongside all their absorbed siblings, with no forwarders.

### Impact

Any installation whose curator/auxiliary model follows the tool schema strictly cannot complete a consolidation. The umbrella-building pass (which is opt-in and explicitly designed to merge narrow skills into class-level skills) is structurally unable to finish its last step. On our instance this held true across three different aux models and two weeks of runs.

### Possible fixes

1. Advertise `absorbed_into` in `SKILL_MANAGE_SCHEMA` (simplest; the ~100 tokens is cheap relative to a silently broken pass), possibly only when the tool context is the curator fork.
2. Or: expose a curator-only schema variant of `skill_manage` for the background review fork.
3. Or: make the guard fail-open-with-warning when the deleted skill's content was demonstrably absorbed this run (existing `_classify_removed_skills` heuristic), instead of refusing unconditionally.
4. For router-dedup gateways: rename the delete-intent parameter so it cannot collide with other keys (e.g. `absorb_target`).

### Related issues

- #29912 (closed) — the fail-open archive bug that motivated `_curator_consolidation_delete_guard`; this is the inverse failure introduced by the fix.
- #76588 — consolidation misclassified as prune when the umbrella is created in the same run.
- #88882 (closed) — curator prompt still instructing direct pruning after deletes were prohibited.

Happy to provide the full `run.json` tool-call trace if useful.}

RelaxJonh:
## Summary

Fixes #97959.

The curator's consolidation pass instructs the review agent to archive absorbed skills with `skill_manage(action="delete", absorbed_into=<umbrella>)`. The handler accepts `absorbed_into`, but the parameter was deliberately omitted from `SKILL_MANAGE_SCHEMA`. Models that strictly follow the tool schema therefore never emit it, every consolidation delete is refused by `_curator_consolidation_delete_guard`, and the pass ends with zero consolidations while reporting success.

## Fix

Add `absorbed_into` as an optional string property in the `delete` operation's schema so schema-compliant models can declare the umbrella skill that absorbs the deleted skill's content.

## Changes

- `tools/skill_manager_tool.py`: Add `absorbed_into` property to `SKILL_MANAGE_SCHEMA` items properties.

## Testing

```sh
# Run curator consolidation — delete ops should now include absorbed_into
hermes curator run --sync --consolidate
# Check logs: consolidated_this_run should be > 0 (not 0)
```

### DIFF
--- a/tools/skill_manager_tool.py
+++ b/tools/skill_manager_tool.py
@@ -2132,6 +2132,14 @@ SKILL_MANAGE_SCHEMA = {
                         "file_content": {
                             "type": "string",
                             "description": "Content for write_file."
+                        },
+                        "absorbed_into": {
+                            "type": "string",
+                            "description": (
+                                "Delete only: name of the umbrella skill that "
+                                "absorbs this skill's content. Required for "
+                                "curator consolidation deletes."
+                            )
                         }
                     },
                     "required": ["name", "action"]


Enough1122:

### PR #98538 — fix(agent): add absorbed_into to skill_manage schema (#97959)

**Summary:** Adds `absorbed_into` (string, delete-only, umbrella skill name) to `SKILL_MANAGE_SCHEMA`. Schema was missing it while handler and `_curator_consolidation_delete_guard` required it, so strict-schema models never emitted it and consolidation deletes were refused with zero consolidations.

- Schema-only change; handler already accepted the field — minimal risk.
- Description notes `Required for curator consolidation deletes` — guides model.

**Verdict:** Minimal schema fix. No blocking issues.

alt-glitch:

Duplicate of #97964. Both advertise `absorbed_into` in the same skill_manage operations schema; #97964 also includes focused regression coverage.

ZeCoL:
Fresh evidence for this PR, and a note about the label parking it.

**The `duplicate` label cites #97964, which does not exist.** `GET /repos/NousResearch/hermes-agent/issues/97964` returns 404, while neighbours 97962, 97963, 97965 and 97966 all resolve — so this is not a deleted-and-renumbered issue, the number was never allocated. The label comes from an automated triage comment ("Duplicate of #97964 ... also includes focused regression coverage") describing a PR nobody can open. @tmaarcxs flagged this on #97959 on 2026-09-05 and it has not been answered. This is currently the only open fix for #97959.

**Reproduced on v0.21.1 (`d15ed444`)**, git install, 194-skill library. Two runs, same config except `auxiliary.curator.model`:

- `gpt-6-astra`/openai-codex: declined to attempt any delete, citing the schema directly — *"It has no `absorbed_into` parameter, so I cannot supply the required forwarding target"*. 5 tool calls, `consolidated_this_run: 0`. Note this variant logs **no** `_curator_consolidation_delete_guard` refusals at all, so it is invisible to anyone grepping for the guard's error string.
- `claude-sonnet-5`/anthropic: 117 tool calls, `consolidated_this_run: 0`, and **zero deletes attempted** — the pass died on `same_tool_failure_halt` after 6 no-op `patch` calls with `old_string == new_string`, two of them carrying explicit probe text. Full detail in my comment on #97959.

So this diff is necessary — both failures trace to the same schema/prompt contradiction. One caveat for reviewers, not an objection to merging: on the second run the schema gap caused the pass to fail *before* the delete guard was ever reached, so this fix alone would not have rescued it. That is a separate defect (the no-op patch refusal is indistinguishable, to the model, from a permission refusal) and belongs in its own issue rather than being bolted onto an 8-line schema fix.

Checks are green on the current head (OSV scan, nix flake check, amd64 + arm64 builds all SUCCESS).

One suggestion on the diff itself, take it or leave it: the description reads *"Required for curator consolidation deletes."* The handler treats `absorbed_into=""` as an explicit prune and a missing value as legacy-accepted, so "required" is true only inside a curator consolidation context. Something like *"Curator consolidation only: the umbrella skill that absorbs this skill's content; pass an empty string to declare a true prune with no forwarding target"* would match `_delete_skill()`'s actual semantics and teach the prune case too, which the current wording leaves undiscoverable.

ZeCoL:
**Correction to my comment above.** I attributed the second run's failure to no-op `patch` calls (`old_string == new_string`). Checking `agent.log` rather than inferring from `run.json`, all 6 refusals are the **read-before-write guard** (#95976) — the model had viewed each target up to 3 times beforehand, the repeat-view dedup returned `content_returned: false` stubs, and the guard refused a file it had just read. Detail posted on #95976.

That run therefore never reached a delete, so it says nothing about `absorbed_into` either way. **Disregard my "necessary but not sufficient" caveat** — it was reasoning from a wrong cause. This PR is the correct and sufficient fix for #97959 as reported, and nothing I found argues against merging it.

What stands from my earlier comment, both independently verified:

1. The `duplicate` label cites **#97964, which does not exist** (404, while 97962/97963/97965/97966 all resolve). That is what is parking the only open fix for a P2.
2. `gpt-6-astra` on v0.21.1 declines to attempt any delete, quoting the schema gap directly — and logs **zero** `_curator_consolidation_delete_guard` refusals while doing so, so this variant is invisible to a log grep for the guard's error string.

The optional wording suggestion on the `absorbed_into` description (covering the `absorbed_into=""` explicit-prune case, which `_delete_skill()` supports but the current text leaves undiscoverable) also stands.

alt-glitch:

Re-triage: removed the `duplicate` label. The cited anchor #97964 is no longer reachable on GitHub (404), so this is the active open fix for #97959. Related: #98605 (curator-side handling of nested `skill_manage` operations, different fix).

PR #98541, opened 2026-08-30
fix: pass normalized context to plugin senders

elkimek:
## Summary

Pass the host-normalized delivery context to plugin `send_message_handler` implementations that opt into the widened contract.

The context includes:

- normalized message text
- filtered media files
- resolved `thread_id`
- `force_document`

Legacy four-argument sync and async plugin handlers remain supported through signature detection.

## Why

Plugin handlers previously received only model-facing `args`. Cron, CLI, and manual callers can invoke `_send_to_platform()` with `args=None` while supplying normalized text/media/thread values directly. A plugin full-request handler could therefore receive an empty request and reject an otherwise valid scheduled send.

## Verification

- `11 passed` in `tests/tools/test_send_message_plugin_extensibility.py`
- Python compilation, Ruff, and `git diff --check` passed
- rebased onto current canonical `main`
- joint core ↔ Fluxer probe preserved a no-args scheduled message, media tuple, thread, and document mode
- independent hostile review before publication: PASS, no actionable P1/P2 findings

## Plugin compatibility

The corresponding Fluxer plugin change is https://github.com/elkimek/hermes-fluxer-plugin/pull/7

The Fluxer plugin gates its widened handler registration on the versioned host capability. Older hosts retain the existing standalone sender rather than enabling an incompatible handler.

### DIFF
--- a/tests/tools/test_send_message_plugin_extensibility.py
+++ b/tests/tools/test_send_message_plugin_extensibility.py
@@ -14,7 +14,7 @@ import pytest
 
 from gateway.config import Platform
 from gateway.platform_registry import PlatformEntry, platform_registry
-from tools.send_message_tool import resolve_send_target, send_message_tool
+from tools.send_message_tool import _send_to_platform, resolve_send_target, send_message_tool
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


Enough1122:

**98541 — fix: pass normalized context to plugin senders**

- `tools/send_message_tool.py:79-83` adds `PLUGIN_SEND_HANDLER_NORMALIZED_CONTEXT=1` as versioned capability for feature-detection by plugins — clear, non-breaking.
- `tools/send_message_tool.py:91-126` adds `_call_plugin_send_handler` that inspects `handler` signature for `normalized` or `**kwargs` before injecting `{"message", "thread_id", "media_files", "force_document"}`. Handles sync vs async via `inspect.isawaitable` correctly. Silent fallback on `TypeError/ValueError` from `signature()` is appropriate for built-ins/C extensions.
- `tools/send_message_tool.py:141-162` dispatches non-builtin platforms early, before Telegram chunk loop, via `platform_registry.get(platform_name).send_message_handler`. Preserves legacy 4-arg contract (`args or {}`) while passing canonical host payload to opted-in handlers. Catches handler exception and returns `{"error": "Plugin send_message handler failed: ..."}` — avoids surfacing stack but leaks exception message (acceptable, same pattern elsewhere).
- `tests/tools/test_send_message_plugin_extensibility.py:19-65` adds `test_plugin_handler_receives_normalized_context_without_model_args` covering `thread_id/media_files/force_document` propagation with `args={}` for cron/CLI origin — directly exercises the lossless path described in PR description.

Non-blocking:
- `inspect.signature` is computed per send; negligible for low-frequency sends but could be cached per handler if hot path grows. **Non-blocking**.
- `VAR_KEYWORD` branch means any `**kwargs` handler receives `normalized` even if unrelated — intentional per comment but worth documenting for plugin authors that `normalized` is reserved. **Non-blocking**.
- Error return shape `{"error": ...}` vs raising differs from earlier chunk-loop paths that raise; callers already handle both, but ensure docs reflect plugin failures surface as error dict not exception. **Non-blocking**.

Verdict: Focused extensibility fix that keeps backward compat while making cron/CLI → plugin media/thread lossless. No blocking issues.

---

PR #70287, opened 2026-07-23
fix(mcp): propagate test failures to CLI exit code

lofibrainwav:
## Summary

`hermes mcp test <name>` previously returned rc=0 even when the connection failed: `cmd_mcp_test` correctly returned 1 on failure, but the dispatcher in `mcp_command()` (mcp_config.py:1094-1096) ignored the return value, and `cmd_mcp` (main.py:14327-14330) further discarded whatever it received. CI, orchestrators, and watchdogs reading `$?` would see success while the real MCP handshake had failed.

Closes the user-visible gap where operators (and automation) could not distinguish a healthy MCP from a dead one based on exit code alone.

## Changes

- `hermes_cli/mcp_config.py`: handler dispatch now propagates integer returns explicitly (None / non-int keeps legacy rc=0 for compatibility). `cmd_mcp_test` itself already returns 1 on connection failure, server not found, or zero tools discovered.
- `hermes_cli/main.py` `cmd_mcp`: when `mcp_command` returns an explicit non-zero rc, `sys.exit(rc)` so the process exit code matches the handler verdict. None / 0 flows through unchanged.

## Validation (live)

```
hermes -p kingdom mcp test neo4j-mcp          # → rc=1
hermes -p kingdom mcp test kg-ingest         # → rc=0
hermes -p kingdom mcp list                   # → rc=0
hermes -p kingdom mcp test __nonexistent__   # → rc=1
hermes -p kingdom mcp bogus                  # → rc=2 (argparse)
```

## Out of scope

- `package-lock.json` and `tests/tools/test_approval.py` changes are reserved for separate PRs.

## Author

승상 Hermes 명장 작성 (현재 MiniMax-M3 모델) · 2026-07-23 PDT

PR #98550, opened 2026-08-30
fix(mcp): add safe application health checks

itmanbrian:
## Summary
- Adds an optional, explicitly configured application-level health check to `hermes mcp test <name>` after MCP discovery.
- Requires an advertised tool, explicit JSON-object arguments, and exact read-only annotations; aliases and result states fail closed on conflicts.
- Keeps ordinary probes/runtime MCP registration discovery-only, documents credential-safe config, and adds regression coverage for typed-object, dict, and generic-Mapping MCP representations.

## Validation
- `TestConfiguredApplicationHealthCheck`: 20 passed
- `scripts/run_tests.sh tests/hermes_cli/test_mcp_config.py tests/hermes_cli/test_mcp_security.py -q`: 103 passed
- ACP regression set: 256 passed using the healthy shared dev venv
- Independent exact-current review: PASS
- Full suite was executed. Its 19 failures across eight unrelated files reproduce exactly at parent `9f275bfb`; the two provider-discovery failures were traced to untracked provider directories in the original worktree and pass in a clean worktree at this head.

## Dependency
- Depends on #70287 (`fix/mcp-cli-exit-code`). GitHub cannot use that external-fork branch as a PR base, so this targets `main` and temporarily includes its two commits.

No runtime tool-registration changes.

### DIFF
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -14330,7 +14330,14 @@ def cmd_plugins(args):
 def cmd_mcp(args):
     from hermes_cli.mcp_config import mcp_command
 
-    mcp_command(args)
+    rc = mcp_command(args)
+    # mcp_command now propagates handler exit codes; honor non-zero so
+    # `hermes mcp test` (and friends) report failure to callers — CI,
+    # orchestrators, watchdogs — instead of silently returning 0.
+    # Handlers that don't override rc continue to flow as None, which
+    # we treat as success (rc=0).
+    if isinstance(rc, int) and rc != 0:
+        sys.exit(rc)
 
 
 def cmd_claw(args):
--- a/hermes_cli/mcp_config.py
+++ b/hermes_cli/mcp_config.py
@@ -719,7 +719,12 @@ def cmd_mcp_list(args=None):
 # ─── hermes mcp test ──────────────────────────────────────────────────────────
 
 def cmd_mcp_test(args):
-    """Test connection to an MCP server."""
+    """Test connection to an MCP server.
+
+    Exit codes:
+      0 — connection succeeded and at least one tool was discovered
+      1 — connection failed, server not found, or zero tools discovered
+    """
     name = args.name
     servers = _get_mcp_servers()
 
@@ -728,7 +733,7 @@ def cmd_mcp_test(args):
         available = list(servers.keys())
         if available:
             _info(f"Available: {', '.join(available)}")
-        return
+        return 1
 
     cfg = servers[name]
     print()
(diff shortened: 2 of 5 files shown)


Enough1122:

**98550 — fix(mcp): add safe application health checks**

- `hermes_cli/mcp_config.py:39-54` adds `_is_json_value` rejecting tuples/non-string keys/NaN/Inf before `health_check.arguments` is accepted — prevents `json.dumps` coercion silently mutating configured args (correct strictness).
- `hermes_cli/mcp_config.py:57-93` adds `_parse_configured_health_check` validating `tool` is non-empty trimmed string and `arguments` is exact JSON object; rejects unknown keys and missing `arguments`. Fails closed with `ValueError` before any network call.
- `hermes_cli/mcp_config.py:118-161` implements `_health_tool_is_read_only` via `_health_alias_value` handling both `readOnlyHint`/`read_only_hint` and dict vs object shapes; `_run_configured_health_check` verifies advertised tool exists, has `readOnlyHint=true`, then calls `session.call_tool`. Error payloads are redacted (`"configured health check RPC failed"` / `"returned an error"`), not logged — good hygiene for sensitive server responses.
- `hermes_cli/mcp_config.py:189,207-212,243-266` gates execution to `run_health_check=True` (only `cmd_mcp_test` sets it), records `details["health_check"]` on success, and shows discovery-only warning otherwise — `tools/list` success no longer implies app credentials verified.
- `hermes_cli/mcp_config.py:221-252,272-284` makes `cmd_mcp_test` return `1` on missing server / exception / zero tools, and `mcp_command`/`hermes_cli/main.py:10-17` propagate that via `sys.exit(rc)` — fixes silent `0` on `hermes mcp test` failures for CI/orchestrators.
- `website/docs/reference/mcp-config-reference.md:66-811` documents opt-in semantics and `readOnlyHint` being server hint not guarantee — accurate caveat.

Non-blocking:
- `_probe_single_server` reads `server._tools` (private attribute) and `_health_value` abstracts Mapping vs attribute access; if upstream renames `_tools` the probe breaks — consider fallback to `getattr(server, "tools", [])`. **Non-blocking**.
- `_is_json_value` rejects top-level non-object `arguments` per spec, but `health_check` could arguably accept empty `{}` only — current allows any JSON object depth, reasonable. **Non-blocking**.
- No explicit timeout on `call_tool` for health probe; inherits `connect_timeout` context but a hung tool could stall `mcp test` longer than expected — consider passing probe-specific timeout. **Non-blocking**.
- `details` out-param mutation is shared across callers; ensure callers don't reuse same dict concurrently. **Non-blocking**.

Verdict: Well-scoped safety improvement with thorough validation, read-only gating, and payload redaction. Exit-code propagation correctly separates discovery vs application health. No blocking issues.

---

Issue #93892, opened 2026-08-24
[Bug]: macOS Desktop — secondary gateway 'retained' flag ignored by pruner → permanent resume/reap spinner; model catalog reads misrouted to ambient gateway


Kkkkkuro:
## Summary

On macOS Desktop (v0.20.5, commit `ddbd928e`), opening a secondary profile's canonical bot chat can enter a **permanent spinner**: the session runtime resumes successfully, then is reclaimed, unbound, and resumed again in a loop with no terminal state and no overall budget. Two distinct defects combine to produce this:

1. **Secondary gateway `retained` flag is not honored by the pruner** (new gap in the #91276/#93361 family, still present on current main).
2. **Model catalog *reads* are misrouted to the ambient gateway** (new call site in the #91684 owner-routing family), which also leaves the model pill spinning forever.

This is NOT a recurrence of #92832 (`activeBotRoute()` no longer exists).

## Root cause 1: retained secondary gateways still get pruned

- `openGatewayForAgent()` marks the selected bot's secondary as `retained=true`, intending foreground/warmup consumers to keep owning the connection after an RPC (`apps/desktop/src/store/gateway.ts:57-58,868-885`).
- `pruneSecondaryGateways()` never checks `entry.retained`; it only keeps the active key, working/needs-input scope, active requests, relay retainers, or temporary activation leases (same file, `1097-1130`).
- The BOTS workspace opens with `keepAllProfilesScope:true`, so the global active profile stays primary/default (`apps/desktop/src/plugins/hermes-bots/plugin.js:5517-5527`; `apps/desktop/src/sdk/index.ts:713-719`). An idle secondary-profile bot is not in the working/attention keep-set (`apps/desktop/src/app/gateway/hooks/use-gateway-boot.ts:709-732`), so its foreground-retained secondary can still be pruned.
- When the secondary WebSocket closes, the backend detaches its runtime and starts the orphan grace reaper (`tui_gateway/ws.py:513-564`; `tui_gateway/server.py:1255-1350,1363-1430`). The just-resumed runtime is reaped again.
[...]

PR #98551, opened 2026-08-30
fix(desktop): route session branches through the parent's owning connection

Zeus-Deus:
## Problem

Branching a session that belongs to one connection while a different connection is active leaves the new chat permanently stuck. The sidebar paints its row — `draft: branch #1`, nested under the parent — and the chat pane spins forever, eventually landing on:

> **Couldn't load this session**
> The connection to this session failed and automatic retries gave up. Check that the gateway is running, then try again.

Retry re-runs the same mis-routed resume, so it never recovers.

There are two independent defects on this path, and fixing only the first leaves the symptom fully intact.

## Defect 1 — the create goes to the wrong backend

`branchStoredSession` and `branchCurrentSession` resolved only the parent's **profile**:

```ts
const profile = sessionProfile ?? stored?.profile

await ensureGatewayProfile(profile)
const { messages } = await getAllSessionMessages(storedSessionId, profile)
```

…then dispatched through the ambient `requestGateway`. A profile name does not identify a backend once more than one connection is configured — several connections routinely expose a profile called `default`. `ensureGatewayProfile` swaps on the profile key alone and has no notion of which connection that profile belongs to, so both the parent transcript read and the `session.create` / `session.branch` RPC land on whichever socket happens to be active.

This is an inconsistency inside the existing code, not a missing capability. `removeSession`, about fifty lines below in the same file, already routes by the full owner:

```ts
const removedOwner: SessionOwnerScope = removed?.connection_id
  ? { connectionId: removed.connection_id, profile: removed.profile || 'default' }
  : profile
```

Registry rows are explicitly tagged for exactly this purpose — `fetchRegistrySessionRows` in `electron/profile-session-routing.ts` sets `session.connection_id = connectionId` on every spliced row so that "opens route correctly".

The transcript read is the quieter half: reading a foreign-owned parent from the active socket returns nothing, so the branch aborts early as "nothing to branch" and no create is attempted at all.

**Fix:** derive the owner with `sessionOwnerRouteFromRow`, activate with `ensureGatewayAgent(connectionId, profile)`, dispatch with `requestGatewayForAgent(...)`, and pass the same owner scope to `getAllSessionMessages`. Both arms of `forkBranch` are covered — `session.branch` for the open chat, `session.create` for a sidebar right-click.

## Defect 2 — the row then lies about who owns it

With the create correctly routed, the symptom persisted, because the row inserted afterwards was untagged.

`upsertOptimisticSession` stamps the row's profile from `$activeGatewayProfile` and **omits `connection_id` entirely** when no owner is passed (`utils.ts:1318-1342`), and it also skips `setSessionOwnerHint`. The branch call site passed no owner, so the child got **neither a row tag nor a hint**.

That matters because `resumeSession`'s owner ladder begins:

```ts
const ownerRoute = capturedOwner || getSessionOwnerHint(storedSessionId)
```

and `forkBranch` calls `resumeSession` without a `capturedOwner` — so the missing hint alone is enough to send the resume to whichever backend is active. The row tag is the second mechanism, consulted by the row-based rungs of the same ladder. The two sibling routed creates in this file already pass their route here; branch simply never did.

**The tile path had the same defect one rung further out.** Branching a session that is *not* the currently-open chat opens a tile rather than resuming, and `SessionTileChrome` resolved its owner from the tile route alone:

```ts
const ownerRoute = sessionTileOwnerRoute(storedSessionId)
```

`openSessionTile` is called for a branch child with no `workspaceScope`, and `session-states.ts` only persists a tile `ownerRoute` in bots mode — so that tile had no owner and its model + composer RPCs fell back to the ambient socket. Its sibling in `session-tile-actions.ts:99-103` already uses the correct tile-route-then-row ladder; this aligns them.

An untagged parent row reproduces the previous ambient behaviour exactly, so single-connection users are unaffected.

## Verification

Reproduced and fixed against **two real gateways**: the desktop's own local backend plus a second `hermes serve` instance registered as a remote connection, with a parent session seeded on the remote.

Driving the **actual sidebar context menu → Branch** in a running dev app:

| | remote backend log | resulting sidebar row |
|---|---|---|
| after fix | `ws closed … messages=11 … detached_sessions=1` | `connection_id: "rigremote"`, stable across 16 polls / 8s |
| before fix | *never saw the request* | no `connection_id` at all |

The child's `model` (`z-ai/glm-5.2`) and `cwd` (`/home/zeus/projects`) in the create response are the **remote's** config, not the local backend's — independent confirmation of which machine answered.

Routing was also checked one level down: running the same `session.create` through `requestGatewayForAgent` vs the ambient `requestGatewayForProfile` shows the remote gateway observing the create only on the owner-routed path.

## Tests

- four call-site regressions in `use-session-actions.test.tsx`: the create rides the owning `(connection, profile)` socket; the transcript read carries the same owner scope; the optimistic row is tagged with the parent's owner; an untagged parent still uses the ambient socket
- three source-contract assertions in `session-tile-owner-route.test.ts` covering the tile owner ladder and the bare-profile narrowing
- `branch-owner-routing.integration.test.ts` mocks nothing inside the router — the real `requestGatewayForAgent` runs against a fake Electron bridge and transport, so a regression that re-collapses a registry route onto the ambient socket fails even when the call-site assertions still pass

Each fix was confirmed to fail before the change: without the row stamp the new test reports `expected undefined to be 'pandora'`.

Full desktop suite green — **6828 tests, 688 files** — with `tsc --noEmit` and eslint clean (no warnings).

## Follow-up in this PR: the socket itself (41c8b3888b0)

Field testing the first two commits against a real remote (branching a remote-owned parent while another connection was active) surfaced a third layer. Routing and row-stamping held — the create landed on the owning backend, the sidebar row was tagged, every individual resume succeeded — but the tile still flickered and eventually latched:

> **Couldn't open this session**
> Session keeps losing its backend runtime right after resuming — retry to resume it again.

That message is the tile's resume-storm breaker (`TILE_RESUME_STORM_LIMIT`, #93892): more than 4 *successful* resumes in 120s. The RPCs were right; the **owning socket** was the casualty. Nothing durable named the owner, so the gateway keep-set never pinned it:

- `openSessionTile` persisted a tile `ownerRoute` only for `workspaceMode === 'bots'`. The branch tile pinned nothing in `openTileGatewayScopes` / `foregroundSessionScopes`; the idle pruner closed the owner socket, the backend orphan-reaped the draft runtime, `session.reclaimed` unbound the tile, resume re-armed and succeeded on a fresh socket that the next recompute closed again — around the loop until the breaker latched. Now a sessions-mode opener that knows the exact owner persists it, and a route-less re-scope (plain sidebar re-open) no longer clobbers it.
- `forkBranch` never called `setSessionOwnerHint` / `holdSessionOwnerUntilForeground` after the create, unlike both sibling routed creates in the same file — leaving the create → tile-publication gap unpinned. Added, mirroring the siblings; the hold retires once the tile's own route covers the scope.
- `resetTileRuntimeBindings` preserved cross-connection runtime bindings only for bot tabs, so an unrelated flapping connection (an SSH source re-dialing) dropped the branch tile's healthy binding on every reconnect — the same storm by another path. Any owner-routed tile is preserved now; the owner's own reconnect still drops the binding for re-resume. This deliberately widens reconnect preservation from bot tabs to all owner-routed tiles — flagged for review.

UX half of the report: the optimistic `draft: branch #N` row survived sidebar refreshes only via `sessionsToKeep()`, which had no rung for open tiles — a sidebar-branch never selects the child, so the next background refresh silently dropped the row and nothing indicated a branch had happened until the first message persisted it. Open tiles' stored ids now join the keep set; the draft row appears immediately and stays.

**Verification** (dev app, remote-owned parent, sibling connection flapping): draft row visible immediately and stable across refreshes; tile carries `{connectionId, profile}`; first turn accepted and completed by the owning backend (`tui prompt accepted` → `turn finished status=complete` in its gateway log); no storm/error card through the full 120s window — before this commit the card latched at ~20s under the same conditions. Each of the three fixes was reverted individually to confirm its regression test fails (RED). Full suite green on this branch: 6765 tests / 684 files, `tsc --noEmit` and eslint clean.

### DIFF
--- a/apps/desktop/src/app/chat/session-tile-owner-route.test.ts
+++ b/apps/desktop/src/app/chat/session-tile-owner-route.test.ts
@@ -11,3 +11,28 @@ describe('SessionTilePane owner-scoped listing', () => {
     expect(source).not.toMatch(/void resolveStoredSession\(storedSessionId\)\s*\n/)
   })
 })
+
+describe('SessionTileChrome owner ladder', () => {
+  // A tile opened without an explicit route (openSessionTile with no
+  // workspaceScope — how a branch child is opened) has no tile ownerRoute, so
+  // the tile route ALONE leaves ownerRoute undefined and requestForSessionProfile
+  // falls back to the ambient socket. The session's own row/hint rung is what
+  // keeps its model + composer RPCs on the backend that owns it.
+  it('falls back to the session row owner when the tile carries no route', () => {
+    expect(source).toContain(
+      'sessionTileOwnerRoute(storedSessionId) ?? knownSessionOwner(ownerLookupSessionRows(), storedSessionId)'
+    )
+  })
+
+  it('does not resolve the chrome owner from the tile route alone', () => {
+    // The pre-fix shape: `const ownerRoute = sessionTileOwnerRoute(storedSessionId)`
+    // with no fallback rung.
+    expect(source).not.toMatch(/const ownerRoute = sessionTileOwnerRoute\(storedSessionId\)\s*\n/)
+  })
+
+  it('narrows a bare profile owner to an object route', () => {
+    // knownSessionOwner may return a bare profile string, which carries no
+    // connection and must not be handed to requestForSessionProfile as a route.
+    expect(source).toContain("typeof resolvedOwner === 'object'")
+  })
+})
--- a/apps/desktop/src/app/chat/session-tile.tsx
+++ b/apps/desktop/src/app/chat/session-tile.tsx
@@ -44,10 +44,12 @@ import {
   $gatewayState,
   $selectedStoredSessionId,
   $sessions,
+  knownSessionOwner,
+  ownerLookupSessionRows,
   sessionMatchesStoredId,
   sessionPinId
 } from '@/store/session'
-import { requestForSessionProfile } from '@/store/session-request-router'
+import { requestForSessionProfile, type SessionOwnerRoute } from '@/store/session-request-router'
 import {
   $sessionStates,
   $sessionTileDelegateRevision,
@@ -157,7 +159,38 @@ function TileChat({
 }) {
   const { gateway, requestGateway } = useGatewayRequest()
   const queryClient = useQueryClient()
-  const ownerRoute = sessionTileOwnerRoute(storedSessionId)
+
+  // Owner ladder, same as useSessionTileActions (session-tile-actions.ts:99-103):
+  // this tile's explicit route first, then the session row's own
+  // (connection, profile) tag — knownSessionOwner also folds in the owner hint.
+  // A tile opened without an explicit route — e.g. a branch child, which
+  // openSessionTile creates with no workspaceScope — has no tile route, so the
+  // row/hint rung is the only thing keeping this tile's model + composer RPCs
+  // on the backend that owns the session instead of the ambient one.
+  //
+  // Resolved on every render (cheap id lookups) so it cannot go stale against
+  // the tile store, the recents/cron/messaging rows, or the hint map. Only the
+  // resulting IDENTITY is memoised, on primitives, because knownSessionOwner
+  // mints a fresh object per call and requestTileGateway below is keyed on it.
+  const resolvedOwner =
+    sessionTileOwnerRoute(storedSessionId) ?? knownSessionOwner(ownerLookupSessionRows(), storedSessionId)
+
+  const ownerConnectionId = resolvedOwner && typeof resolvedOwner === 'object' ? resolvedOwner.connectionId : ''
+  const ownerProfile = resolvedOwner && typeof resolvedOwner === 'object' ? resolvedOwner.profile : ''
+  const ownerTargetProfile = resolvedOwner && typeof resolvedOwner === 'object' ? resolvedOwner.targetProfile : undefined
+
+  // A bare profile string carries no connection and is not a usable route here.
+  const ownerRoute = useMemo<SessionOwnerRoute | undefined>(
+    () =>
+      ownerConnectionId
+        ? {
+            connectionId: ownerConnectionId,
+            profile: ownerProfile,
+            ...(ownerTargetProfile ? { targetProfile: ownerTargetProfile } : {})
+          }
+        : undefined,
+    [ownerConnectionId, ownerProfile, ownerTargetProfile]
+  )
 
   const requestTileGateway = useCallback(
     <T,>(method: string, params?: Record<string, unknown>, timeoutMs?: number, signal?: AbortSignal): Promise<T> =>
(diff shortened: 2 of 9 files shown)


Enough1122:

**98551 — fix(desktop): route session branches through the parent's owning connection**

- `apps/desktop/src/app/session/hooks/use-session-actions/index.ts:269-320` extends `forkBranch` with optional `ownerRoute: SessionOwnerRoute`. When present it `await ensureGatewayAgent(connectionId, profile)` and routes via `requestGatewayForAgent(connectionId, profile, method, params)`; otherwise preserves legacy `ensureGatewayProfile(profile)` + `requestGateway`. Mirrors `removeSession` ownership contract and fixes profile-name collision across connections (`pandora` vs `other-box` both `default`).
- `apps/desktop/src/app/session/hooks/use-session-actions/index.ts:340-360,369-401` applies same routing to transcript fetch: `getAllSessionMessages(storedId, ownerRoute ?? profile)` plus `sessionOwnerRouteFromRow(stored)` extraction. Prevents "nothing to branch" abort when parent lives on foreign backend and ambient read returns empty.
- `apps/desktop/src/app/session/hooks/use-session-actions.test.tsx:148-256` adds three unit cases: connection-tagged create rides `requestGatewayForAgent('pandora','default',...)` not ambient, transcript read uses `(connectionId,profile)` route, and untagged parent keeps ambient socket — no behavior change for single-backend users.
- `apps/desktop/src/app/session/hooks/branch-owner-routing.integration.test.ts:1-133` adds real-router integration test with `FakeHermesGateway` + fake `window.hermesDesktop` registry exposing same profile on two connections. Asserts `dialed` contains correct `connectionId` and `sent` URL contains `pandora`, and that two same-named profiles use distinct sockets — catches profile-only routing regression that mocked unit tests would miss.

Non-blocking:
- `ensureGatewayAgent` also stamps `$activeGatewayProfile` per earlier comment; verify `upsertOptimisticSession` still sees correct profile after connection swap — existing branch tests pass but worth manual multi-backend smoke. **Non-blocking**.
- `ownerRoute` is optional and threaded via `forkBranch(branchMessages, null, ..., profile, undefined, ownerRoute)` — the `undefined` placeholder for `branchCount` is fragile if signature grows; consider options-object param in future. **Non-blocking**.
- Integration test uses `vi.resetModules()` + `window.hermesDesktop` mock; ensure cleanup between suites doesn't leak dialed state (cleared in `beforeEach`). **Non-blocking**.

Verdict: Correct owner-aware routing with strong regression coverage (unit + real router). Untagged path unchanged, multi-connection ambiguity resolved.

---

OutThisLife:
Superseded by #100879, which consolidates the full branch-load path \u2014 owner routing, seeded-child persistence, route navigation, create coalescing, and self-describing project rows. Your commits are cherry-picked there with authorship preserved.\n\nThanks for this \u2014 the owner-routing diagnosis and the two-gateway reproduction were the backbone of the fix.

PR #98553, opened 2026-08-30
fix(cron): anchor croniter to the configured IANA timezone

jflemieux:
## What does this PR do?

`compute_next_run` (`cron/jobs.py`) hands croniter a tz-aware base time, and croniter works in
the base's UTC *offset* rather than in its zone. A fixed offset cannot track DST, so a
`0 9 * * *` job in America/Toronto fires at **08:00 local on spring-forward day and 10:00 on
fall-back day**, and stays an hour off for the rest of each season. That is the live bug this
PR fixes.

The fix renders the base as the *configured* IANA zone's naive wall clock (`get_timezone()`)
before handing it to croniter, then re-attaches the zone to the result, so the wall-clock hour
stays correct on every calendar day including DST boundaries. It falls back to the base's own
zone when no timezone is configured — behaviour-identical to `main` in that case (verified
below).

One correction to this PR's original description: the UTC-stored `last_run_at` half it also
claimed (a 09:00 job firing at 09:00 UTC) is **already fixed on `main`**, and was fixed before
this branch existed — `605ba4adea` ("interpret naive timestamps as local time") makes
`_ensure_aware`/`_parse_aware` normalize a stored instant into the configured zone before it
reaches croniter. This PR does not claim that half, and its test for it has been dropped.

The branch also carries a merge of current `main`, which had refactored this same branch of
`compute_next_run` (base hoisted into one `_parse_aware` call, croniter call collapsed to a
single line). The merge keeps main's structure and re-applies the anchoring on top of main's
hoisted `base_time`; this PR's own `base_time` block is dropped as superseded. No other file
conflicted, and `git diff origin/main..HEAD` is exactly these two files.

## Related Issue

No existing issue — self-contained fix with regression tests.

Reproduction on `main` (c7c86a7dc6), zone configured the way production configures it
(`HERMES_TIMEZONE=America/Toronto`), expr `0 9 * * *`:

```
last=2026-03-07T09:00:00-05:00 -> 2026-03-08T08:00:00-04:00   # 08:00 EDT, expected 09:00
last=2026-10-31T09:00:00-04:00 -> 2026-11-01T10:00:00-05:00   # 10:00 EST, expected 09:00
```

The added test file is red on `main` and green after the fix.

## Type of Change

## Changes Made

- `cron/jobs.py` — in the `cron` branch of `compute_next_run`, render the base as
  `get_timezone()`'s naive wall clock for croniter and re-attach the zone to its result
  (applied to the `base_time` that `main` now hoists above the kind branches). Falls back to
  the base's own zone when nothing is configured.
- `tests/cron/test_compute_next_run_dst.py` — two invariant tests: the spring-forward and
  fall-back days both keep the 09:00 local hour, and a 365-day walk across both transitions
  never drifts. Both are red on unfixed `main` with the shipped symptom.

## How to Test

1. `scripts/run_tests.sh tests/cron/test_compute_next_run_dst.py` — on `main` both tests fail
   (the spring-forward day yields 08:00 local); after the fix, 2 passed.
2. `scripts/run_tests.sh tests/cron/test_compute_next_run_dst.py tests/cron/test_compute_next_run_last_run_at.py` — 3 passed
3. `scripts/run_tests.sh tests/cron/` — 1248 passed, 1 skipped (windows_only, covered by the tests-os Windows lane)
4. Manual: with `HERMES_TIMEZONE=America/Toronto` the reproduction above returns 09:00 local
   for both transition days, and with no timezone configured the output is byte-identical to
   `main` across cron/interval/once, naive, UTC-stored and malformed `last_run_at` inputs.

### Code

- [x] I've run `scripts/run_tests.sh tests/cron/` and all tests pass (1248 passed, 1 skipped)
- [x] I've tested on my platform: Linux (CachyOS / Arch rolling, Python 3.11)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (inline comment + test docstrings cover the change)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (pure datetime/zoneinfo logic, stdlib-only)


Enough1122:

**98553 — fix(cron): anchor croniter to the configured IANA timezone**

- `cron/jobs.py:9,30-34` imports `get_timezone()` and computes `zone = get_timezone() or base_time.tzinfo`, then `base_wall = base_time.astimezone(zone).replace(tzinfo=None)` + `cron.get_next(...).replace(tzinfo=zone)`. Correctly works around `croniter` 6.0 ignoring `tzinfo` and using offset-as-working-offset, which previously caused UTC-stored `last_run_at (+00:00)` to fire at 09:00 UTC instead of 09:00 Toronto, and DST days to drift to 08:00/10:00.
- `cron/jobs.py:17-29` comment accurately explains wall-clock vs UTC offset distinction and references morning-routine 09:00 `America/Toronto` — aligns with reported strand.
- `tests/cron/test_compute_next_run_dst_and_utc_last_run.py:82-127` adds four regression cases: UTC-stored last_run stays 09:00 local, spring-forward, fall-back, and 365-day walk asserting `(hour,minute)==(9,0)` every day — covers both reported bugs.

Non-blocking:
- Fallback to `base_time.tzinfo` when `get_timezone()` is `None` preserves prior behavior for unconfigured hosts, but if `base_time` itself was naive (shouldn't be per `_ensure_aware`) `zone` would be `None` and `.replace(tzinfo=None)` would still succeed — ensure `compute_next_run` callers always pass aware `now`. **Non-blocking**.
- Ambiguous wall-clock hour during fall-back (1:30 occurs twice) will resolve to first occurrence via naive croniter — acceptable for 09:00 jobs but document if sub-hour cron ever used at 01:xx. **Non-blocking**.
- Test monkeypatches `cron.jobs.get_timezone`; production `get_timezone` may cache ZoneInfo — no issue, just note test isolation. **Non-blocking**.

Verdict: Minimal, well-explained fix for wall-clock correctness with strong DST coverage. No blocking issues.

---

teknium1:
Salvaged into #109532 with your commits cherry-picked (authorship preserved), plus a fold-aware strictly-after guard for the DST fall-back hour (ported from an upstream fix in another agent runtime) and the elapsed-duration fix from #109414. Will auto-close when that lands. Thanks!

PR #98555, opened 2026-08-30
agent: expose first stream chunk timestamp (TTFB) in post_api_request hook

overtoneblue:
## Motivation
`post_api_request` already carries `started_at`, `ended_at`, and `api_duration` for every API call, but nothing separates time-to-first-token from generation time. Observers cannot compute TTFB or true output throughput (`output_tokens / (ended_at - first_chunk_at)`) — tokens-per-second reported over the whole call understates real generation speed whenever TTFB is non-trivial (reasoning models, provider queueing, slow TTFB providers).

## Change
The streaming path (`interruptible_streaming_api_call`) already measures `first_chunk_at` in its per-attempt diagnostics (`request_client_holder["diag"]`, see `agent/stream_diag.py`) — but only for failure telemetry; the value is discarded on success. This PR surfaces it:

1. Stash the latest successful attempt's `first_chunk_at` on the agent (`agent._last_api_first_chunk_at`) at stream completion (`agent/chat_completion_helpers.py`).
2. Reset it at the top of each API attempt (`agent/conversation_loop.py`, next to `_reset_stream_delivery_tracking()`), so a stale value can never leak into the next call's payload.
3. Forward it as `first_chunk_at` in the existing `post_api_request` hook payload, alongside `started_at`/`ended_at`.
4. Extend the synthetic `post_api_request` test payload in `hermes_cli/hooks.py` to include the timestamp fields.

No new instrumentation in the stream hot path — consumer-side TTFB = `first_chunk_at - started_at`.

## Compatibility
Additive kwargs only; existing `post_api_request` subscribers and shell hooks ignore unknown fields. `first_chunk_at` is `None` for non-streamed responses and failed attempts.

## Notes
- `first_chunk_at` is the first provider chunk received, which for reasoning models may precede the first visible content delta.
- Follows the existing stash-on-agent convention (`agent._last_turn_usage` in the same loop).


Enough1122:

**98555 — agent: expose first stream chunk timestamp (TTFB) in post_api_request hook**

- `agent/chat_completion_helpers.py:9-16` after successful `interruptible_streaming_api_call`, propagates `request_client_holder["diag"]["first_chunk_at"]` onto `agent._last_api_first_chunk_at` as float — bridges per-attempt stream diagnostics (set on first chunk in `chat_completion_helpers`) to conversation loop without threading return values.
- `agent/conversation_loop.py:31` resets `agent._last_api_first_chunk_at = None` at start of each attempt in `run_conversation` retry loop — prevents stale value leaking from previous attempt into `post_api_request` (explicitly noted in comment).
- `agent/conversation_loop.py:39-46` forwards `first_chunk_at=getattr(agent, "_last_api_first_chunk_at", None)` into hook payload alongside `started_at/ended_at/api_duration` — enables `TTFB = first_chunk_at - started_at`, `None` when non-streamed or no chunk.
- `hermes_cli/hooks.py:58-60` updates `_cmd_list` example payload to include `started_at/ended_at/first_chunk_at` — docs stay in sync.

Non-blocking:
- `_diag_last` is read after `result["response"] is not None` check; on empty/error paths `first_chunk_at` stays `None` — correct. Ensure `request_client_holder` is always dict (it is per `interruptible_streaming_api_call` contract). **Non-blocking**.
- Attribute is set on `agent` dynamically (`_last_api_first_chunk_at`) without type stub — consider adding `Optional[float]` annotation to `Agent` class for discoverability. **Non-blocking**.
- If streaming is disabled mid-attempt, `diag` may lack `first_chunk_at`; `None` propagation is handled by hook consumers but document null case. **Non-blocking**.

Verdict: Small, well-isolated observability hook with proper per-attempt reset and no functional change to streaming. No blocking issues.

---

andrexibiza (commented):
Reviewed exact head `4ba91eedf8cd9bdca65d7f88a9fd30041bab63fa`.

**Blocking**

- This head has no executed validation. The CI, Nix flake, and Docker workflows all completed as `action_required`, and the combined status contains no check contexts. Until those workflows are approved/rerun and green on this exact SHA, the behavior is unverified and the PR is not merge-ready.
- Add regression coverage for the state-transfer contract: a successful streamed response emits the latest attempt’s non-null `first_chunk_at`; non-streamed, failed, and partial-stream paths emit `None`; and an outer retry/reset cannot leak a prior attempt’s timestamp. The value crosses a helper-local diagnostic into mutable agent state and later hook dispatch, so stale-state behavior needs executable proof.

I traced `stream_diag.py` → `interruptible_streaming_api_call()` → `_last_api_first_chunk_at` reset → `post_api_request` dispatch. The changed control flow is internally consistent; I found no separate correctness defect in the patch.

teknium1:
Merged via #100425 (rebase-merge, your commit's authorship preserved — merge SHA e17276c7b4, now on main). Regression tests requested in review were added on top (tests/run_agent/test_first_chunk_at_hook.py). Thanks @overtoneblue!

PR #98559, opened 2026-08-30
fix(cron): deliver secondary-profile cron jobs via their own adapters

gordianschosser:
## Problem

With `multiplex_profiles`, one gateway process serves several Hermes profiles, each with its own bot identity (Telegram/Discord/…). Cron jobs are ticked per profile store, but **every cron delivery went out through the default profile's adapter map** (`runner.adapters`). The secondary profiles' own adapters live in `runner._profile_adapters` and were never handed to the scheduler.

Visible symptom in a multi-bot setup: a job fired from a secondary profile store delivered its DM to the user **from the default profile's bot identity** — or failed outright where the default bot token cannot resolve the target chat. Job execution was correctly scoped per profile (skills, memory, store), only the delivery identity was wrong.

## Root cause

`InProcessCronScheduler._start_multiplex` ticks every profile home with the single shared `adapters` map. `gateway/run.py` built `cron_start_kwargs` with only `adapters` — secondary adapter maps (`runner._profile_adapters`) never reached the scheduler.

## Fix

- `cron/scheduler_provider.py`: `start()` and `_start_multiplex()` accept an optional `profile_adapters` (profile name → that profile's live adapter map). The multiplex tick loop extracts the profile name from each `profile_homes` entry and, for secondary profiles, ticks with `profile_adapters.get(profile_name) or {}` — **never falling back to the default profile's adapters**. If a secondary profile's adapter map is absent or empty (e.g. its bot has not connected yet during startup), that profile simply does not deliver this tick instead of sending through the wrong bot identity. The default profile (or a `None` name) keeps ticking with the shared `adapters` map, which it owns.
- `gateway/run.py`: pass `runner._profile_adapters` and the default profile name into `cron_start_kwargs` when secondary adapters exist.
- `tests/cron/test_scheduler_provider.py`: regression tests asserting (a) the default profile ticks with the shared adapters, (b) a connected secondary profile ticks with its own adapters, (c) a secondary profile without an adapter entry gets an empty map — no fallback to the default bot.

No behavior change for single-profile or non-multiplex installations: the parameters default to `None`, the default profile path is unchanged, and the gateway only injects `profile_adapters` when secondary adapters actually exist. Verified: existing multiplex tests stay green.

## Tests

- New: `tests/cron/test_scheduler_provider.py` — four policy tests (default-shared / connected-own / empty-no-fallback / missing-no-fallback) plus the existing per-profile adapter regression test, extended for the `default_profile` parameter (RED before fix, GREEN after)
- `tests/cron/`: 1052 passed, 10 skipped (1 pre-existing, approval-setting-dependent failure — fails identically on clean main)
- `tests/gateway/test_multiplex_profile_authz.py`: 6 passed

## Notes

The no-fallback policy for secondary profiles mirrors the safety rule from #73363: an unconnected secondary profile must stay silent rather than deliver through the wrong bot. Ported manually onto current main because `scheduler_provider.py` gained `fire_claimed` / `_existing_profile_homes` / misfire backstop since the original patch.

### DIFF
--- a/cron/scheduler_provider.py
+++ b/cron/scheduler_provider.py
@@ -558,6 +558,7 @@ class InProcessCronScheduler(CronScheduler):
         interval=60,
         can_dispatch=None,
         profile_homes=None,
+        profile_adapters=None,
     ):
         import logging
         from cron.scheduler import tick as cron_tick
@@ -582,6 +583,7 @@ class InProcessCronScheduler(CronScheduler):
                 stop_event,
                 profile_homes=profile_homes,
                 adapters=adapters,
+                profile_adapters=profile_adapters,
                 loop=loop,
                 interval=interval,
                 can_dispatch=can_dispatch,
@@ -653,6 +655,7 @@ class InProcessCronScheduler(CronScheduler):
         *,
         profile_homes,
         adapters=None,
+        profile_adapters=None,
         loop=None,
         interval=60,
         can_dispatch=None,
@@ -664,6 +667,14 @@ class InProcessCronScheduler(CronScheduler):
         agent execution to that profile's home — mirroring how
         ``_profile_runtime_scope`` scopes the multiplexed inbound path and
         ``web_server.py`` scopes per-profile cron API calls.
+
+        ``profile_adapters`` maps profile name → that profile's live adapter
+        map (``{Platform: adapter}``). When provided, each profile ticks with
+        ITS OWN adapters so cron deliveries issued from a secondary profile's
+        store ride that profile's bot identity instead of the default
+        profile's. Fall back to the shared ``adapters`` map for profiles not
+        present in the map (default profile, or a gateway that never started
+        secondary adapters).
         """
         import logging
         from cron.scheduler import tick as cron_tick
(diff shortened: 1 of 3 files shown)


alt-glitch:

Related: #73363 is the earlier focused multiplex adapter repair. It deliberately supplies an empty adapter map when a secondary profile is absent or unconnected; this patch falls back to the default adapters, which can still send through the wrong bot during startup. Please reconcile that policy before selecting a patch.

gordianschosser:
Reconciled — thanks for the pointer to #73363. This PR now adopts the same safety policy:

- **No default-adapter fallback for secondary profiles.** The multiplex tick loop now ticks a secondary profile with `profile_adapters.get(profile_name) or {}` — an absent or empty adapter map means that profile simply does not deliver that tick, instead of sending through the default profile's bot during startup.
- **`default_profile` parameter** added so the default profile (and `None` names) keeps the shared `adapters` map it owns.
- **Regression tests updated/added** (default-shared / connected-own / empty-no-fallback / missing-no-fallback), RED → GREEN. `tests/cron/`: 1052 passed, 10 skipped (1 pre-existing, approval-setting-dependent, fails identically on clean main); `tests/gateway/test_multiplex_profile_authz.py`: 6 passed.

PR description rewritten to describe the current behavior and is intentionally generic (no install-specific profile names).

teknium1:
Closing after #99375 (merged, commit 9eb832aad7) landed the delivery-identity fix via earlier submitters @MuhammadUsamaMX (#73363) and @ghosty93 (#92526), whose versions carried tests. Same root cause, later submission. Thanks for the correct analysis.

PR #98563, opened 2026-08-30
fix(ci): fail aggregate gate on cancelled jobs

### LINKED ISSUE #98557: all-checks-pass reports success when detect is cancelled and every lane skips
## Summary

`all-checks-pass` in `.github/workflows/ci.yaml` treats `cancelled` the same as `skipped`, so a run in which the `detect` job is cancelled reports **green** while having executed no tests at all.

## Mechanism

The evaluation step fails only on the literal string `failure`:

```python
failed = [name for name, info in needs.items() if info['result'] == 'failure']
icon = '✅' if result in ('success', 'skipped') else '❌'
if failed:
    sys.exit(1)
print('All checks passed (or were skipped)')
```

Every lane-gated job is conditioned on a `detect` output, e.g.:

```yaml
tests:
  needs: detect
  if: needs.detect.outputs.python == 'true'
```

When `detect` is cancelled its outputs are empty, so each of those conditions evaluates false and the job resolves to `skipped` rather than `failure`. `all-checks-pass` then sees a set of `skipped` results, finds nothing equal to `failure`, and exits 0.

## Observed

A run where `detect` was cancelled ~6 seconds in, during its `actions/checkout` step. Result: `Python tests`, `OS-specific tests`, `JS & TS checks`, `Rust tests`, `Installer tests`, `Docs Site`, `Supply-chain scan`, `Review label gate` and the rest all `skipped` — and `All required checks pass` reported **success**.

## Why this matters

Treating `skipped` as a pass is correct and deliberate: it is what makes lane-gating work, and skipping the Rust lane on a docs-only PR is the intended behaviour.

`cancelled` is a different thing. `skipped` is a *decision* — the classifier ran and concluded the lane was irrelevant. `cancelled` is an *absence* — nothing concluded anything. Collapsing them means a cancelled `detect` silently disables the entire matrix and still reports green, because the aggregate cannot distinguish "we decided not to test this" from "we never found out".

`all-checks-pass` is the natural required-status-check target for branch protection, which is where this becomes load-bearing: a mid-run cancellation would satisfy a required check over a completely untested PR. Runs are cancellable routinely — `concurrency.cancel-in-progress` is set to true for `pull_request` events in the same file.

## Suggested fix

Count `cancelled` (and the empty-string result) as failures:

```python
BAD = {'failure', 'cancelled', ''}
failed = [name for name, info in needs.items() if info['result'] in BAD]
```

Optionally special-case `detect` itself: if it did not succeed, fail the aggregate outright, since every downstream decision depends on its outputs.

huklaa:
Fixes #98557.

## Problem

`all-checks-pass` only treated the literal `failure` result as blocking. If `detect` was cancelled, every lane depending on its outputs could skip and the aggregate gate still reported success despite no test decision having been made.

## Fix

Keep intentional `skipped` lanes accepted, but treat every dependency result other than `success` or `skipped` as blocking. This covers `cancelled` and avoids silently accepting incomplete/unknown result states.

The aggregate output/icons remain unchanged; only the exit decision and nearby workflow comment are tightened.

## Validation

- refreshed onto current upstream `main` (`21b2095d00a98b8ad7b5c60b10587619c852cdb8`);
- branch is exactly one focused commit ahead and zero commits behind;
- final duplicate search still finds no other open PR covering #98557 / this cancelled aggregate-gate path;
- fresh hosted CI, Docker, and Nix checks are running on head `68ca8e5b9ea2ec6be9fb4eecafef67ff2cd5896f`.

Focused gate semantics remain:
- success + intentional skipped lanes -> pass
- cancelled dependency -> block
- failed dependency -> block
- empty/incomplete result -> block

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -205,8 +205,8 @@ jobs:
 
   # ─────────────────────────────────────────────────────────────────────
   # Gate: runs after everything. ``if: always()`` ensures it reports a
-  # status even when some deps were skipped. Only actual ``failure``
-  # results cause it to fail; ``skipped`` is treated as success.
+  # status even when some deps were skipped. Intentional ``skipped`` results
+  # are accepted; cancelled or otherwise incomplete dependencies must fail.
   #
   # Branch protection should require ONLY this check.
   #
@@ -256,13 +256,17 @@ jobs:
           print(f'needs-json={json.dumps(compact)}')
           with open('$GITHUB_OUTPUT', 'a') as f:
               f.write(f'needs-json={json.dumps(compact)}\n')
-          failed = [name for name, info in needs.items() if info['result'] == 'failure']
+          blocked = [
+              name
+              for name, info in needs.items()
+              if info['result'] not in ('success', 'skipped')
+          ]
           for name, info in sorted(needs.items()):
               result = info['result']
               icon = '✅' if result in ('success', 'skipped') else '❌'
               print(f'{icon} {name}: {result}')
-          if failed:
-              print(f'::error::{len(failed)} job(s) failed: {\", \".join(failed)}')
+          if blocked:
+              print(f'::error::{len(blocked)} job(s) did not succeed or skip: {\", \".join(blocked)}')
               sys.exit(1)
           print('All checks passed (or were skipped)')
           "


Enough1122:

**98563 — fix(ci): fail aggregate gate on cancelled jobs**

- `.github/workflows/ci.yaml:8-12` updates gate comment to "Intentional `skipped` results are accepted; cancelled or otherwise incomplete dependencies must fail" — matches new logic.
- `.github/workflows/ci.yaml:21-34` changes aggregation from `info['result']=='failure'` to `info['result'] not in ('success','skipped')`, so `cancelled`, `timed_out`, and other non-success states now block `All checks passed`. Icon logic updated consistently (`✅` only for `success/skipped`). Previously a cancelled job was treated as success, hiding infra flakes.
- `ci.yaml:42` whitespace newline fix at EOF — trivial.

Non-blocking:
- Branch protection should require only this gate (per comment) — ensure `skipped` remains intentional via job `if:` conditions, not accidental infra skips; current allowlist is correct (common pattern is `success/skipped` only). **Non-blocking**.
- Error message `"did not succeed or skip"` accurately reflects new semantics vs old `"failed"` — good signal for triage. **Non-blocking**.

Verdict: Tight, correct hardening of required-checks gate. No blocking issues.

---

PR #35670, opened 2026-05-31
fix(bedrock): batch QoL — raise max_tokens, strip status bar prefix, context table, max_iterations

JiaDe-Wu:
Salvage of 4 low-risk community PRs onto current main. Each addresses a real user pain point, all are small and self-contained.

### 1. Raise default max_tokens from 4096 to 16384 (ref #20619 by @liorfranko)

The Converse API path hardcoded `max_tokens=4096`. Claude Opus 4.6+ supports 32K output, and even Sonnet regularly produces 8K+ tool-call responses. Users hit silent truncation (`finish_reason='length'`) on any non-trivial agent task.

### 2. Strip Bedrock regional/vendor prefix from status bar (ref #19849 by @mmcclean-aws)

`us.anthropic.claude-sonnet-4-6` is useful as a model ID but ugly in the TUI status bar. Now displays as `claude-sonnet-4-6`. Strips known regional prefixes (`us.`, `global.`, `eu.`, `ap.`) and vendor prefixes (`anthropic.`, `amazon.`, `meta.`, etc.).

### 3. Align context-window table with current Anthropic docs (ref #24059 by @patrick-muller)

Adds missing entries: Opus 4.7/4.8, Nova Premier (1M), Llama 4 Maverick (1M), Pixtral Large, DeepSeek R1, and 2P models (Kimi K2, GLM 4, MiniMax M2). Without correct context lengths, the compression threshold calculates wrong and either compresses too early or too late.

### 4. Add bedrock_converse branch in handle_max_iterations (ref #27544 by @EloquentBrush0x)

`handle_max_iterations()` dispatched `codex_responses` and `anthropic_messages` to their native paths, but `bedrock_converse` fell through to the generic OpenAI-client path (which has no Bedrock credentials). The iteration-limit summary now routes through `call_converse()` directly.

## Testing

118 bedrock_adapter tests passing. All changes are in the Bedrock-specific code path — no impact on other providers.

## Credit

Original authors credited in commit message. This PR rebases their work onto current main (all 4 original PRs are 50-200+ commits stale).

PR #27544, opened 2026-05-17
fix(agent): add bedrock_converse branch in handle_max_iterations

EloquentBrush0x:
## Problem

`handle_max_iterations` in `agent/chat_completion_helpers.py` dispatches `codex_responses` and `anthropic_messages` to their respective native API paths, but falls through to the generic OpenAI-client path for `bedrock_converse`:

```python
if agent.api_mode == "codex_responses":
    ...
else:
    # builds OpenAI summary_kwargs
    if agent.api_mode == "anthropic_messages":
        ...
    else:
        # ← bedrock lands here and calls:
        agent._ensure_primary_openai_client().chat.completions.create(...)
```

Bedrock uses boto3 directly — there is no OpenAI client. The call fails, the outer `except Exception` catches it, and the user sees:

```
I reached the maximum iterations (90) but couldn't summarize. Error: <boto3/httpx error>
```

instead of a real summary. This affects every Bedrock-mode agent that hits `max_iterations`.

## Root Cause

Split Logic Bug — `bedrock_converse` has explicit branches in every other dispatch function in this file:

| Function | `bedrock_converse` handled |
|---|---|
| `build_api_kwargs` (L261) | ✅ |
| `interruptible_api_call` (L110) | ✅ |
| `interruptible_streaming_api_call` (L1165) | ✅ |
| `handle_max_iterations` (L967+) | ❌ before this PR |

## Fix

Add an `elif agent.api_mode == "bedrock_converse":` branch in both the first attempt and the retry path. The branch follows the same pattern already used in the three functions above:

1. Build kwargs via `_build_api_kwargs` (which calls `BedrockTransport.build_kwargs` and embeds `__bedrock_region__` / `__bedrock_converse__` sentinels)
2. Pop `toolConfig` — the Bedrock Converse API key for tool definitions (confirmed in `bedrock_adapter.py` and `transports/bedrock.py`)
3. Call `_interruptible_api_call`, which dispatches to `boto3 client.converse()`
4. Normalize via `BedrockTransport.normalize_response`, which handles the already-normalized `SimpleNamespace` returned by `_interruptible_api_call`

## Test Plan

- [ ] Bedrock-mode agent that hits `max_iterations` now receives a real text summary instead of an error message
- [ ] `toolConfig` is absent from the boto3 `converse()` call so no tool-calling occurs in the summary turn
- [ ] `codex_responses` and `anthropic_messages` paths are unchanged
[...]

PR #70132, opened 2026-07-23
fix(agent): sanitize iteration-limit summary failures

plcunha:
## Summary

Sanitize failures from the final no-tools summary request when Hermes reaches `max_iterations`.

The old catch-all branch interpolated `str(error)` into both the user-facing response and warning log. Structured provider errors can contain quota-plan/reset metadata, while arbitrary transport exceptions can contain credentials or tokens.

This PR preserves the summary attempt and returns stable, actionable guidance without exposing the raw exception.

## Root cause

`handle_max_iterations()` treated the final summary failure as ordinary display text:

```python
f"... couldn't summarize. Error: {str(e)}"
```

That crossed an exception-safety boundary: provider/SDK internals became user-visible and log-visible.

## What changed

- Reuse the central `classify_api_error()` taxonomy.
- Return tailored guidance for rate/usage limits, billing/credit limits, authentication failures, and generic failures.
- State truthfully that work completed before the limit remains visible in the conversation.
- Never interpolate the raw exception into the response or warning log.
- Log only exception type, classified reason, and HTTP status.
- Preserve the successful summary path, retry behavior, Relay logical-call identity, and finalization outcome from current `main`.

## Security properties

Regression tests assert that responses and logs do not contain:

- raw provider payloads;
- plan/reset metadata;
- arbitrary exception text;
- API-key-shaped private values.

The generic response remains actionable without claiming that a summary was produced.

## Current-main port

The branch was rebuilt on current `main` (`12b1f0f83d`). The surrounding summary path had evolved since the original patch, so the port deliberately preserves the newer Relay lifecycle contract (`complete_logical_call` in `finally`) and its retry identity tests. No current-main tests were deleted.

## Verification

```text
python -m pytest -q \
  tests/run_agent/test_run_agent.py \
  tests/agent/test_max_iterations_summary_failure.py
# 263 passed

python -m pytest -q \
  tests/agent/test_error_classifier.py \
  tests/agent/test_chat_completion_helpers_provider_sort.py \
  tests/agent/test_turn_finalizer_iteration_limit_exit.py \
  tests/agent/test_turn_finalizer_final_response_persistence.py \
  tests/agent/test_turn_finalizer_cleanup_guard.py \
  tests/agent/test_turn_finalizer_interrupt_alternation.py
# 102 passed

python -m ruff check \
  agent/chat_completion_helpers.py \
  tests/agent/test_max_iterations_summary_failure.py \
  tests/run_agent/test_run_agent.py
# All checks passed

[...]

PR #98564, opened 2026-08-30
fix(bedrock): iteration-limit summary falls through to the OpenAI client on bedrock_converse

JiaDe-Wu:
> **Rebased onto `01ae7a5668` on 2026-09-06 — the fix shape changed, because `main` did.**
> `main` replaced the `if/elif` ladder this PR used to patch with a dispatch table
> (`_SUMMARY_ATTEMPT_BUILDERS`, `agent/chat_completion_helpers.py:2282`) and that table has no
> `bedrock_converse` key, so the defect survived the refactor unchanged while this diff came to
> aim at code that no longer exists. It is now **two commits**: a pure move of the 213-line
> iteration-limit-summary unit into `agent/chat_completion_helpers_iteration_limit.py`
> (@andrexibiza's bounded-owner gate, `AGENTS.md:245`), then one builder registered under
> `"bedrock_converse"` plus the regression tests. `+489/-211`; the 3,569-line owner drops to 3,363.
> Details, the two boundary calls I made, and the mutation results are in
> [this comment](https://github.com/NousResearch/hermes-agent/pull/98564#issuecomment-5559369665).
> The ladder quoted under **Problem** below is the pre-refactor shape, kept because it is what the
> defect looked like when it was found; the mechanism and the user-visible symptom are unchanged.

> **Prior art — #27544 got here first.** The triage bot is right that this duplicates
> [#27544](https://github.com/NousResearch/hermes-agent/pull/27544) (@EloquentBrush0x, 2026-05-17).
> They found the bug and patched the same two branches three and a half months before I did.
> I've added a `Co-authored-by:` trailer so the credit lands with them regardless of which PR merges.
> The concrete deltas in this one are listed in [Relationship to #27544](#relationship-to-27544) at
> the bottom — if the maintainers would rather take theirs, say so and I'll close this and port the
> tests onto their branch instead.

## Problem

`handle_max_iterations` dispatches on `api_mode` but only names two of them (pre-refactor shape; the table at `:2282` has the same gap):

```python
if agent.api_mode == "codex_responses":       # ...
elif agent.api_mode == "anthropic_messages":  # ...
else:
    summary_client = agent._ensure_primary_openai_client(
        reason="iteration_limit_summary"
    )
```

`bedrock_converse` falls into the `else`. Bedrock talks to boto3 and has no OpenAI client, so the call raises, the wrapping `except Exception` at the end of the function swallows it, and the user gets:

```
I reached the maximum iterations (N) but couldn't summarize. Error: ...
```

**Every time.** The iteration-limit summary has never worked on Bedrock. The same `else` appears twice — the initial summary and its retry — so both are affected, and the retry cannot rescue the first.

## The part that isn't just a missing feature

The `else` branch does not fail fast. When the agent holds live OpenAI-compatible credentials — the common case, since a `base_url`/`api_key` pair is usually configured even by Bedrock users — `_ensure_primary_openai_client()` returns the existing client and the summary request is actually **sent**: the full conversation, addressed to `agent.base_url`, carrying a Bedrock `modelId`.

Measured against unpatched `main` (`66666f6e2e`), driving the real `handle_max_iterations` with `api_mode="bedrock_converse"`:

```
api_mode           : bedrock_converse
openai client used : True
base_url it targets: https://openrouter.ai/api/v1
model sent         : us.anthropic.claude-sonnet-4-5-20250929-v1:0
conversation in it : True
user got           : I reached the maximum iterations (...) but couldn't summarize. Error: 401 ...
```

So a Bedrock session's conversation is transmitted to a non-Bedrock endpoint before failing there. It fails because no such model exists at that provider — but the request body has already left the process. For a user who chose Bedrock specifically to keep traffic inside their AWS account, that is the wrong destination.

## Fix

Add the missing branch to **both** paths, routing through `_dispatch_nonstreaming_api_request` rather than calling `converse()` here:

```python
elif agent.api_mode == "bedrock_converse":
    final_response = _bedrock_iteration_summary(
        agent, api_messages, _managed_summary_call, retry_count=0
    )
```

Three deliberate choices:

- **Reuse the one dispatch point.** `_dispatch_nonstreaming_api_request`'s docstring says per-api_mode dispatch "lives in exactly one place", and its bedrock branch already carries cachePoint-rejection recovery (#97281) and stale-connection client eviction. A hand-rolled `converse()` call here would silently lose both.
- **Reuse the existing kwargs builder.** `build_api_kwargs` already has a `bedrock_converse` branch; the summary path was simply never taught to call it. No new region/guardrail resolution logic.
- **`tools_for_api=[]`.** The summary prompt says "without calling any more tools", so sending `toolConfig` invites exactly what it forbids. This mirrors the `codex_kwargs.pop("tools", None)` two branches up. (`build_converse_kwargs` gates on `if tools:`, so an empty list emits no `toolConfig` at all.)

`make_client` is passed a function that raises. The bedrock branch documents that it never calls it; if a refactor ever does, this fails loudly instead of handing `None` to an SDK.

## Tests

`tests/agent/test_bedrock_iteration_limit_summary.py` — 7 tests. They drive the **real** `BedrockTransport` and the **real** `normalize_converse_response`, stubbing only the boto3 client, so they cover the actual wiring (kwargs shape, sentinel-key stripping, response normalization) rather than a mock's shape.

| test | pins |
|---|---|
| `test_summary_is_produced_over_converse` | the regression: real summary, not the error string |
| `test_no_openai_client_is_ever_built` | the misrouting above — makes touching that client fatal |
| `test_converse_kwargs_are_well_formed` | `__bedrock_converse__`/`__bedrock_region__` stripped before boto3; no `role="system"` message |
| `test_summary_offers_no_tools` | no `toolConfig`, even with `agent.tools` populated |
| `test_empty_first_reply_falls_through_to_the_retry_path` | the **retry** path is fixed too, and also avoids the OpenAI client |
| `test_summary_is_appended_to_history` | the summary becomes the turn's assistant message |
| `test_make_client_hook_is_fatal_not_silent` | the never-called contract |

**All 7 fail on unpatched `main`.** In fairness the last one fails only because the helper does not exist yet, so the meaningful count is 6/7 failing on behaviour.

```
tests/agent/test_bedrock_iteration_limit_summary.py .......  7 passed
```

Full bedrock + summary + transport suites on this branch: **498 passed**. One unrelated pre-existing failure, `test_auxiliary_transport_autodetect.py::test_resolve_provider_client_kimi_coding_wraps_anthropic`, which I verified fails identically on `66666f6e2e` with this branch's source file reverted.

`ruff check` clean on both files.

## Provenance

This bug was item 4 of my stale PR #35670, which I closed on 2026-08-30 because its other three items were wrong or obsolete — one of them would have regressed `main`. This is that one real finding, extracted, rewritten to use the existing dispatch and kwargs builder instead of importing `call_converse` directly as the old PR did, and given the regression test it never had.

+239 / −0. No behaviour change for any other `api_mode`.

**Correction (2026-09-02), prompted by @andrexibiza's review.** Two things in this body were
wrong and are fixed above and below.

The item number was 3; it is 4. More importantly, #35670's heading for that very item reads
*"### 4. Add bedrock_converse branch in handle_max_iterations (ref #27544 by
@EloquentBrush0x)"* — the reference to #27544 was already in my own PR body, in my own words.
#35670's last edit is `2026-08-30T12:40:09Z`; the triage bot's duplicate call on this PR is
`2026-08-30T12:48:00Z`, eight minutes later. So I had recorded the prior art before anything
flagged it, and then wrote this PR as though I had not. The credit was owed from the start
rather than discovered for me, and the closing paragraph below has been corrected to say so.

**Supersession, stated explicitly.** #27544 is the direct implementation predecessor for this
defect class and is still open. This PR is offered not as a duplicate but as the superseding
convergence object: the same two branches, plus `_managed_summary_call` registration, the
canonical `_dispatch_nonstreaming_api_request` route, the wrong-destination witness, and 7
regression tests. The contributor chain stays visible either way — @EloquentBrush0x is
`Co-authored-by:` on the commit and named at the top of this body — and #27544 should be
retired only after this one is accepted, not before.

## Relationship to #27544

#27544 is the same fix, first. Its diff:

```python
elif agent.api_mode == "bedrock_converse":
    bedrock_kwargs = agent._build_api_kwargs(api_messages)
    bedrock_kwargs.pop("toolConfig", None)
    summary_response = agent._interruptible_api_call(bedrock_kwargs)
```

Three differences, in descending order of how much they matter:

1. **`_managed_summary_call` wrapping.** #27544 calls `_interruptible_api_call` directly, which does not go through `relay_llm`. The summary therefore isn't registered with `call_role="iteration_summary"`, `api_request_id`, `retry_count`, or `defer_logical_completion=True` — the metadata the anthropic and OpenAI-wire branches both supply. In fairness the `codex_responses` branch doesn't wrap either, so this is "matches the majority of branches", not "fixes a second bug".
2. **Tests.** #27544 has none; this adds 7, all of which fail on unpatched `main`.
3. **`tools_for_api=[]` vs `pop("toolConfig")`.** Equivalent in effect — `build_converse_kwargs` gates on `if tools:` — so not building it is just tidier than building and discarding it.

**Corrected:** an earlier version of this line said I did not know about #27544 when I wrote this. That is not true. #35670's own heading for the extracted item names #27544 and @EloquentBrush0x, and it said so before the triage bot flagged the duplicate here — see the correction note under [Provenance](#provenance). What actually happened is that I extracted the item from my own stale PR without re-reading the attribution I had written into it, and did not search open PRs either. Both were mine to get right.


alt-glitch:

Duplicate of #27544: both PRs add the missing `bedrock_converse` branches for initial and retry iteration-limit summaries, preventing the OpenAI-client fallback.

andrexibiza (commented):
Reviewed exact head `7b770842188643b44519394cdd34b89f90b462f2` against current `main@66666f6e2eca0ae883195a34c66131985ea7dd06`. This is a one-commit, direct-parent object. I read the full diff, the non-streaming dispatch owner, Bedrock transport normalization, exact-head checks, existing review surface, and the surrounding iteration-limit/Bedrock lineage.

The mechanism itself is strong. Routing the final summary through `_dispatch_nonstreaming_api_request()` is the right ownership boundary: Bedrock stays on its boto3 path, the summary inherits the current cachePoint-rejection and stale-client recovery, and `tools_for_api=[]` keeps the terminal turn genuinely toolless. The tests have useful teeth rather than just asserting a mock shape: they poison OpenAI-client construction, drive the real `BedrockTransport`/normalizer, check sentinel stripping and system-message conversion, prove `toolConfig` is absent even when tools exist, exercise the empty-first-response retry, and verify the summary is persisted into history.

Exact-head hosted acceptance is also clean: [CI 33312094104](https://github.com/NousResearch/hermes-agent/actions/runs/33312094104), [Docker 33312093706](https://github.com/NousResearch/hermes-agent/actions/runs/33312093706), and [Nix 33312093701](https://github.com/NousResearch/hermes-agent/actions/runs/33312093701) all succeeded; `check-attribution` and the aggregate `All required checks pass` are green. There is only one submitted commit, so the whole submitted chain has an exact-object receipt.

Two landing gates remain:

1. **Attribution / supersession provenance.** The current body says this bug was item 3 of #35670. The live #35670 record shows the max-iterations Bedrock slice was **item 4**, and it explicitly records that slice as derived from [#27544](https://github.com/NousResearch/hermes-agent/pull/27544) by `EloquentBrush0x`. #27544 is still open and is the direct implementation predecessor for the same defect class. #98564 is materially stronger and current — canonical dispatcher, both summary attempts, wrong-destination witness, and real regression coverage — so I would treat this PR as the **superseding convergence object**, not as a duplicate. Please correct the body to preserve #27544 / `EloquentBrush0x` credit and state that supersession relationship explicitly. There is no reason to rewrite the existing commit solely for that; preserve the contributor chain and retire the stale competing object only after this one is accepted.

2. **2K / bounded-owner gate.** `agent/chat_completion_helpers.py` is already far beyond the repository line ceiling: this diff adds the new helper around the `2948` hunk and adds the retry branch around `3304`. The new code is coherent, but another provider/lifecycle unit should not accrete into this owner. Please extract the iteration-limit-summary orchestration into a bounded `<2K` module while retaining the original namespace/patch seam for callers/tests that depend on `agent.chat_completion_helpers.handle_max_iterations`. That extraction is also the natural convergence point for open [#70132](https://github.com/NousResearch/hermes-agent/pull/70132), which independently hardens the same handler against leaking raw provider exception text. Whichever lands second should recompose on the first rather than resolving the collision by dropping either security property.

Interlocks I checked: #70132 is complementary failure-sanitization work on the same handler; [#34476](https://github.com/NousResearch/hermes-agent/pull/34476) is adjacent Bedrock request-option work and should compose naturally because this PR reuses `_build_api_kwargs`; [#38282](https://github.com/NousResearch/hermes-agent/pull/38282) is adjacent grace-call lifecycle work and does not eliminate this fallback boundary; #35670 is the historical intermediate salvage object; #27544 is the direct predecessor whose contributor credit should remain visible.

Once those two hard gates are closed, I do not see a runtime-correctness blocker in the Bedrock routing change itself. The wrong-destination regression and reuse of the canonical dispatcher are exactly the right shape. Nice work turning the old stale slice into a much stronger current-main fix. 🚀

JiaDe-Wu:
The duplicate call is correct — thanks. #27544 (@EloquentBrush0x, 2026-05-17) patched the same two branches three and a half months before this, and I didn't know about it: this came out of extracting the one real item from my own stale #35670, not from searching open PRs first, which is the search I should have run.

I've added a `Co-authored-by:` trailer so the credit lands with them whichever PR merges, rewritten the body to lead with the prior art, and commented on #27544 with the misrouting evidence and two suggestions for refreshing it.

Deltas here, honestly ranked: (1) the call is wrapped in `_managed_summary_call`, so the summary is registered with relay_llm like the anthropic and OpenAI branches — #27544's `_interruptible_api_call` isn't, though the codex branch isn't either, so it's consistency rather than a second bug; (2) 7 regression tests, which #27544 has none of; (3) `tools_for_api=[]` instead of building `toolConfig` and popping it, which is equivalent in effect.

If the maintainers prefer #27544, I'll close this and port the tests onto their branch — that outcome is fine by me.

JiaDe-Wu:
Thanks — this is a genuinely useful review, and gate 1 turned out to be worse than you flagged.

## Gate 1 — attribution / supersession: closed, plus a correction I owed

You were right that the item number was wrong: #35670's slice was **4**, not 3. Fixed in the body.

The part you didn't have to hand: #35670's heading for that item reads

> `### 4. Add bedrock_converse branch in handle_max_iterations (ref #27544 by @EloquentBrush0x)`

So the reference to #27544 was in my own PR body, in my own words. And the timestamps rule out the reading that I added it after being told — #35670's last edit is `2026-08-30T12:40:09Z`, the triage bot's duplicate call on this PR is `2026-08-30T12:48:00Z`, eight minutes later.

Which means my earlier comment here, saying I did not know about #27544, was false. What actually happened is that I extracted the item from my own stale PR without re-reading the attribution I had written into it. The credit was owed from the start, not discovered for me. I've corrected both the body and that closing paragraph rather than quietly editing the number, since the wrong version is what you and @alt-glitch read.

Supersession is now stated explicitly in the body: #27544 is the direct implementation predecessor, still open, and this is offered as the superseding convergence object rather than a duplicate — with @EloquentBrush0x kept as `Co-authored-by:` on the commit and named at the top, and #27544 retired only after this is accepted, not before.

## Gate 2 — the bounded-owner extraction: I want to get the target right before I move 5.7k lines

I'm not pushing back on the principle — `agent/chat_completion_helpers.py` is **5536 lines on this branch, 5688 on `main`**, and accreting another provider lifecycle unit into it is clearly the wrong direction. Two honest problems before I start cutting, though.

**I can't find the ceiling written down.** I searched `CLAUDE.md`, `AGENTS.md`, `docs/`, `.github/workflows/`, `pyproject.toml` and the ruff config on current `main` for a file-length rule and found nothing — the only `2000` hits are `max_line_length` for `read_file`'s line-numbered view and a table in an RFC. So I don't know whether "<2K" is a hard repo gate I should treat as binding, or your standard as reviewer. Either is fine by me; I'd just rather comply with the actual shape than guess and produce a refactor you then have to re-scope. If it is written down somewhere I missed, point me at it and I'll follow it exactly.

**Sequencing, because you named the collision yourself.** You noted #70132 independently hardens this same handler, and that whichever lands second should recompose on the first. An extraction is the most conflict-generating change possible here — it moves the very lines #70132 edits. So doing it *inside this PR* maximises the chance that one of us has to redo work, and it couples a 4-line correctness fix (a Bedrock session's conversation currently leaves the process addressed to a non-Bedrock endpoint) to a structural refactor of a shared 5.7k-line module.

My preference, and I'll do whichever you pick:

- **(a)** Land the Bedrock branch as-is, then I open the extraction as its own PR — `agent/iteration_limit_summary.py`, per-`api_mode` summary orchestration, with `agent.chat_completion_helpers.handle_max_iterations` kept as a re-export so every existing caller and `mock.patch` seam is untouched. That PR is then the clean convergence point for #70132 and for retiring #27544, and it can be reviewed as a refactor with no behaviour change to argue about.
- **(b)** Do the extraction here. I'll take the conflict risk with #70132 and coordinate on their PR; just confirm the target module name and boundary you want, and whether the `<2K` figure is binding.

I lean (a) purely because it keeps a P2 data-egress fix from waiting on a refactor, but you have the better view of how this repo likes to sequence these. Say which and I'll get on it.

One correction to my own earlier framing while I'm here: I described delta 1 (`_managed_summary_call`) as consistency rather than a second bug, and I stand by that — the `codex_responses` branch doesn't wrap either, so it's "matches the majority", not a fix.

JiaDe-Wu:
@andrexibiza thank you for that review — it is the most useful one I have had on this repo. Status on both gates, plus a finding that changes the shape of the fix.

## Gate 1 — attribution / supersession: I believe this is closed

Done in the body before this comment: the item number is corrected to **4**, #35670's own heading for that item (`### 4. Add bedrock_converse branch in handle_max_iterations (ref #27544 by @EloquentBrush0x)`) is quoted, the opening block credits #27544 / @EloquentBrush0x as prior art, there is a **Supersession, stated explicitly** section, a `Relationship to #27544` section with the concrete deltas, and the commit carries a `Co-authored-by:` trailer. I also corrected an earlier line of mine that had claimed I did not know about #27544 — untrue, and the correction says so.

If anything there still reads wrong to you, name it and I will fix it.

## A finding: `main` refactored the ladder away, and the defect survived it

The `if/elif` ladder this PR patches **no longer exists on `main`**. It is now a dispatch table, which is why this PR reads `CONFLICTING`:

```python
# agent/chat_completion_helpers.py:2282
_SUMMARY_ATTEMPT_BUILDERS = {"codex_responses": _codex_summary_attempt,
                             "anthropic_messages": _anthropic_summary_attempt}
```

`handle_max_iterations` is now 49 lines (`:2285-2333`) and does `_SUMMARY_ATTEMPT_BUILDERS.get(agent.api_mode, _chat_summary_attempt)`.

**The table has no `bedrock_converse` key**, so a Bedrock agent still falls through to `_chat_summary_attempt` and still constructs the OpenAI client for the terminal turn. The defect is unchanged; only its shape moved. So the rebased fix is a `_bedrock_summary_attempt` builder plus one table entry — a fraction of the `+243` sitting here now, which is aimed at code that is gone.

## Gate 2 — the extraction, measured

The unit is cleaner to lift than I expected. It is **contiguous, `:2121-2333`, 213 lines, 11 symbols**:

| | |
|---|---|
| constants | `_SUMMARY_FOREIGN_MESSAGE_KEYS` `:2121`, `_EMPTY_SUMMARY_RESPONSE` `:2123`, `_SUMMARY_ATTEMPT_BUILDERS` `:2282` |
| message build | `_iteration_summary_api_messages` `:2126`, `_iteration_summary_chat_kwargs` `:2189` |
| call plumbing | `_managed_summary_call` `:2178`, `_summary_text` `:2247` |
| builders | `_codex_summary_attempt` `:2251`, `_anthropic_summary_attempt` `:2259`, `_chat_summary_attempt` `:2271` |
| entry point | `handle_max_iterations` `:2285` |

Every one of the ten private symbols appears exactly **twice** in the file — its definition and its single use inside this unit. Nothing else in the 3,569-line owner consumes them, so this is a move rather than an untangling.

The seam matters to more than the tests: `handle_max_iterations` is imported by `agent/turn_finalizer.py`, `run_agent.py`, and six test modules (`test_api_content_sidecar.py`, `test_turn_finalizer_cleanup_guard.py`, `test_turn_finalizer_final_response_persistence.py`, `test_turn_finalizer_iteration_limit_exit.py`, `tests/run_agent/test_run_agent.py`, `test_verification_continuation_budget.py`), so `agent.chat_completion_helpers.handle_max_iterations` has to keep resolving and keep being patchable.

**Plan, two commits, per `AGENTS.md:245`** (*"split along `<stem>_<topic>` FIRST, in its own commit"*):

1. **Pure move** — those 213 lines to `agent/chat_completion_helpers_iteration_limit.py`, with the facade re-exporting so the seam above is byte-compatible. No behaviour change; existing tests must pass untouched, which is the proof the move is a move.
2. **Behaviour** — `_bedrock_summary_attempt` registered under `"bedrock_converse"`, plus the regression tests from this PR ported onto the new home, including the wrong-destination witness that poisons OpenAI-client construction.

Two questions before I move 213 lines, since the boundary is your call:

- Is `agent/chat_completion_helpers_iteration_limit.py` the name you want, or would you rather it read `chat_completion_helpers_summary.py`? The unit is the iteration-limit summary specifically, not summaries generally.
- Should the three existing builders move with it? They are only reachable through `_SUMMARY_ATTEMPT_BUILDERS`, so leaving them behind would split the table from its handlers — but it also means commit 1 touches `codex_responses` and `anthropic_messages` territory, not just Bedrock's.

On #70132: agreed it converges here. If it lands first I will recompose commit 2 on top of its sanitization rather than around it; if this lands first the extracted module is where its guard belongs, and I will say so on that PR.

JiaDe-Wu:
Rebased onto `01ae7a5668`. Both of @andrexibiza's gates are closed, and the diff is a lot smaller than what it replaces.

## Two commits, in the order `AGENTS.md:245` asks for

**1. `refactor(agent): the iteration-limit summary lives in its own module`** — the pure move. 213 lines, eleven symbols, out of the 3,569-line owner into `agent/chat_completion_helpers_iteration_limit.py` (244 lines). No behaviour change.

**2. `fix(bedrock): give bedrock_converse a real iteration-limit summary`** — `_bedrock_summary_attempt` registered under `"bedrock_converse"`, plus the seven regression tests.

Net `+489/-211` across three files; the owner drops from 3,569 to 3,363 lines.

## Gate 1 — attribution

Closed, and I verified the trailer is actually on the commit rather than only claimed in the body: `Co-authored-by: EloquentBrush0x <283442588+EloquentBrush0x@users.noreply.github.com>`, carried through the rebase onto commit 2. The body's supersession section, the corrected item number (4, not 3), and the `Relationship to #27544` section are unchanged.

## Gate 2 — the extraction, and the two boundary calls I made

You asked me to name them; I went ahead so there was something concrete to reject.

**Name: `chat_completion_helpers_iteration_limit.py`**, not `_summary.py`. The unit is the iteration-limit summary specifically — `handle_max_iterations` is its only entry point — and `_summary` would invite the compaction and title-generation summaries to accrete into it later, which is the failure mode the split exists to prevent.

**The three existing builders moved with it.** They are only reachable through `_SUMMARY_ATTEMPT_BUILDERS`, so leaving them behind would put a dispatch table in one module and its handlers in another. You are right that this means commit 1 touches `codex_responses` and `anthropic_messages` territory rather than only Bedrock's — that is the cost, and I think a table separated from its handlers is the more expensive one. Easy to reverse if you disagree.

**Two helpers stayed behind** because the main loop uses them too — `_provider_preferences_for_agent` (`:1477`) and `_merge_nous_portal_messages_extra_body` (`:1403`). They are imported *inside* the two functions that need them, which matches this unit's own existing idiom for `relay_llm`, `MAX_ITERATIONS_SUMMARY_REQUEST` and `get_provider_profile`, and leaves the new module with no module-level import of its former owner — so there is no cycle to be careful about later.

**The seam is intact and I checked it rather than assuming.** `run_agent.py:1329` binds the entry point through `agent.lazy_forward.forward("agent.chat_completion_helpers", "handle_max_iterations")`, and `lazy_forward`'s docstring is explicit: *"Targets are resolved on every call, so `patch("<module>.<name>")` in tests still intercepts."* The facade re-exports the same function object (`facade.handle_max_iterations is unit.handle_max_iterations` → `True`), so `lazy_attr` resolves it and the six test modules that patch or import it by that path are unaffected.

## The fix got much smaller, because `main` moved

The `if/elif` ladder this PR used to patch is gone — `:2282` is now a table:

```python
_SUMMARY_ATTEMPT_BUILDERS = {"codex_responses": _codex_summary_attempt,
                             "anthropic_messages": _anthropic_summary_attempt}
```

with no `bedrock_converse` key, so the defect survived the refactor unchanged. The fix is one builder and one entry instead of a fourth branch duplicated across the initial and retry paths — and because the retry is the same `attempt` closure the loop already calls twice, the two paths cannot diverge the way two hand-written branches could.

## Evidence

**The wrong-destination claim is now demonstrated, not inferred.** With the fixture pointed at a real endpoint, the pre-fix run came back:

```
I reached the maximum iterations (...) but couldn't summarize.
Error: Error code: 401 - {'error': {'message': 'Missing Authentication header', 'code': 401}}
```

A 401 from `agent.base_url` means the conversation had actually left the process for a non-Bedrock host. The committed fixture points at `127.0.0.1:9` instead, so the suite cannot reach the network either way — matching what `test_api_content_sidecar.py:451` already does.

**Mutation:** removing the one table entry takes **6 of the 7** tests down in 2.2s with `Failed to get summary response: Connection error.` — offline and deterministic. The 7th (`test_make_client_hook_is_fatal_not_silent`) only pins `_bedrock_summary_no_client`'s contract, which that mutation leaves in place, so its staying green is correct rather than a coverage gap. I would rather say that than report 7/7.

**Regression:** 257 tests across the seam and every module that touches the moved symbols pass unmodified — that is the proof the move is a move.

## One thing worth knowing about that regression run

Running those modules together stalls indefinitely, and it is **not** this PR: the same set stalls 3 runs out of 3 on pristine `01ae7a5668` with my commits absent. I narrowed it to a two-test reproducer, got the stack (an `AIAgent()` construction spinning in `plugin_compat.manifest_path`'s `Path.resolve()` at 109% CPU), and filed it as #104287 with two hypotheses I tested and refuted. I mention it here only so that a stalled full-suite run on this branch is not read as this diff's doing.

andrexibiza (commented):
Re-reviewed exact head `131e1c50512b1d64e061eb4624ffc3eabfe3c84f` and the full two-commit chain against the live base.

The two implementation gates from my prior review are closed.

- **Attribution / supersession:** the fix commit now actually carries `Co-authored-by: EloquentBrush0x <283442588+EloquentBrush0x@users.noreply.github.com>`, the body preserves #27544 as the direct predecessor and explicit supersession chain, and the historical item number is corrected to 4.
- **Bounded owner:** the iteration-limit unit is now isolated in `agent/chat_completion_helpers_iteration_limit.py` (291 lines at this head), while `handle_max_iterations` remains re-exported through `agent.chat_completion_helpers` for the existing lazy-forward / patch seam. Keeping the builder table with its handlers is the coherent boundary.
- **Bedrock routing:** `_bedrock_summary_attempt()` builds through the existing `_build_api_kwargs(..., tools_for_api=[])` path and dispatches through `_dispatch_nonstreaming_api_request()`. I checked the live dispatcher at this head: its `bedrock_converse` branch goes directly to `_bedrock_converse_call(..., stream=False)` and does not invoke `make_client`, so `_bedrock_summary_no_client` is a valid fail-loud guard rather than a latent runtime trap. The retry also reuses the same provider-specific attempt closure, so the initial and retry paths cannot drift apart again.

I do not see a new source-level runtime-correctness blocker in these updates.

There is, however, one **landing blocker: the verification object is not green**.

Exact-head [CI 34030977420](https://github.com/NousResearch/hermes-agent/actions/runs/34030977420) is `cancelled`. `Detect affected areas` was cancelled during checkout; Python tests, Python lints, contributor checks, OS-specific tests, and the other dependent jobs were consequently skipped. The aggregate `All required checks pass` result is not an acceptance receipt for this head when the selector job never completed. [Docker 34030976999](https://github.com/NousResearch/hermes-agent/actions/runs/34030976999) and [Nix 34030976993](https://github.com/NousResearch/hermes-agent/actions/runs/34030976993) are green, but they do not substitute for the cancelled CI workflow.

The branch was also rebuilt on `01ae7a5668ce0fa2efca524a4567cacdd0786c95`; live `main` is now `089bb32886c8c18f7fa20182c7bf8826d6935ac5`, 34 commits ahead, and `agent/chat_completion_helpers.py` changed in that interval. The new lazy import of `_provider_preferences_for_agent` should compose with the intervening provider-routing change, but that composition has not been verified on the current base.

Finally, commit 1, `6350a4384861d158401aa75ea9121db20b908e79`, currently has no hosted workflow/status receipt at all. This review's acceptance standard is every retained commit green, not only the tip.

So the code gates are closed, but this is not ready to land yet: rebase onto current `main`, then get fully green exact-head CI on the resulting object and a green hosted receipt for every retained commit. Once those receipts exist, I am clear on the Bedrock/extraction changes.

JiaDe-Wu:
@andrexibiza rebased onto current `main`, and both receipts you asked for now exist. Nothing about the code changed except what the rebase forced; the details of that are below.

## The receipts

Your standard was *"fully green exact-head CI on the resulting object and a green hosted receipt for every retained commit."* Both commits, both green:

| retained commit | CI | Docker | Nix |
|---|---|---|---|
| **1** `6281ddcfd99eca09568be039e4b426f2a9088dce` | [34217447927](https://github.com/NousResearch/hermes-agent/actions/runs/34217447927) ✅ | [34217446837](https://github.com/NousResearch/hermes-agent/actions/runs/34217446837) ✅ | [34217446845](https://github.com/NousResearch/hermes-agent/actions/runs/34217446845) ✅ |
| **2 (head)** `c2cd7699727ecac7d02d8d56e09a6f73dbd51c6a` | [34218010341](https://github.com/NousResearch/hermes-agent/actions/runs/34218010341) ✅ | [34218009503](https://github.com/NousResearch/hermes-agent/actions/runs/34218009503) ✅ | [34218009530](https://github.com/NousResearch/hermes-agent/actions/runs/34218009530) ✅ |

At the head: `Detect affected areas` **SUCCESS** — that is the job whose cancellation invalidated the last receipt, so it is the one I checked first — plus `Python tests / Run tests`, `Python tests / e2e`, `Python lints / ruff enforcement (blocking)`, `Python lints / Windows footguns (blocking)`, `Python lints / ruff + ty diff`, `OS-specific tests / macOS-only tests`, `OS-specific tests / Windows-only tests`, `Check contributors / check-attribution`, and the aggregate `All required checks pass`. 24 SUCCESS, 12 SKIPPED, 1 NEUTRAL (`osv-scanner`, its baseline state), **0 failures, nothing pending**. `mergeable` is back to `MERGEABLE`.

**How commit 1 got a receipt, stated plainly** — commit 1 has never had one because it has never been a PR head. So I pushed the rebased commit 1 alone as this PR's head, let the `pull_request` workflows complete green on that exact object, and then pushed the tip on top. That second push was a **fast-forward, not a force**, so commit 1 is retained unmodified in the chain that is up now — the object with the receipt is the object in the branch. Both rows above are `pull_request` runs on this PR against the same base. If you would rather see this shape done a different way, say so and I will redo it.

## What the rebase had to compose with

Base moved `01ae7a5668` → `3dcf0deb29`, **709 commits**. The rebase produced exactly one conflict hunk, and it was the expected one: main's copy of the unit against my facade note.

**main changed the unit while it sat here, and I carried the change rather than dropping it.** Diffing the unit region base-vs-main, the only substantive change is five lines added to `_iteration_summary_api_messages`:

```python
        # Mirror of the transport's role-qualified strip: ``name`` is
        # schema-foreign on tool results only (strict providers reject with
        # "contains item with unknown key name"); it stays on user/assistant.
        if api_msg.get("role") == "tool":
            api_msg.pop("name", None)
```

Those five lines are in the moved module. A resolution that took my side of the hunk wholesale would have silently reverted them, which is the failure mode a 709-commit rebase invites, so I want it on the record that they are there.

**The composition you flagged as unverified is verified.** You wrote that *"the new lazy import of `_provider_preferences_for_agent` should compose with the intervening provider-routing change, but that composition has not been verified on the current base."* Both helpers I deliberately left behind still exist on the current owner — `_provider_preferences_for_agent` at `:413` and `_merge_nous_portal_messages_extra_body` at `:445` — and the in-function imports resolve. The seam itself I checked by identity rather than by inspection:

```
facade.handle_max_iterations is unit.handle_max_iterations  -> True
lazy_forward.forward("agent.chat_completion_helpers", "handle_max_iterations")  -> resolves
```

**One of the six seam test modules moved in the interval.** `test_verification_continuation_budget.py` is now under `tests/run_agent/`, not `tests/agent/`. It passes (7 passed); I mention it only because the module list in my earlier comment is now stale by one path.

## Re-verified on the new base

- **The 7 regression tests: 7 passed.**
- **Mutation still load-bearing.** Removing the one `"bedrock_converse"` table entry takes **6 of 7** down. The 7th is still `test_make_client_hook_is_fatal_not_silent`, which pins `_bedrock_summary_no_client`'s contract that the mutation leaves in place — same reading as before, and I would still rather report 6/7 than claim 7/7.
- **Seam modules pass individually:** `test_api_content_sidecar.py` 27, `test_turn_finalizer_cleanup_guard.py` 4, `test_turn_finalizer_final_response_persistence.py` 8, `test_turn_finalizer_iteration_limit_exit.py` 9, `test_verification_continuation_budget.py` 7, `test_run_agent.py` 283. (`test_run_agent.py` also has one failure in *my* venv — `TestAnthropicInterruptHandler::test_interruptible_anthropic_interrupt_never_closes_shared_client`, an `ImportError` for the `anthropic` package I have not installed. It fails identically on pristine `3dcf0deb29`, and CI's `Run tests` is green, so it is my box.)
- **Lints.** `ruff check .` → *All checks passed*, and CI's blocking ruff job agrees. `ty`: the owner goes **40 → 39** diagnostics, net **−1**, and that one diagnostic is the one that moved with its own line into the new module — so the extraction is diagnostic-neutral, not diagnostic-shifting. The new test module adds 4 `unresolved-attribute` reports for attributes set dynamically on `AIAgent` (`api_mode`, `model`, `_bedrock_region`, `_cached_system_prompt`); `agent.model = ` appears in **61** existing test modules, so that is this repo's test idiom rather than something new, and I would rather match it than add casts no other test file has.

## Unchanged

Owner drops **3337 → 3126** lines; the unit is 296 lines in `agent/chat_completion_helpers_iteration_limit.py`. Net `+494/-216` across three files. `Co-authored-by: EloquentBrush0x <283442588+EloquentBrush0x@users.noreply.github.com>` is on commit 2, carried through this rebase (verified on the commit object, not just the body). The body's supersession section, the corrected item number (4), and the `Relationship to #27544` section are all untouched. Both boundary calls from last time stand: the module name stays `_iteration_limit` rather than `_summary`, and the three existing builders stay with their table.

On #70132: still the natural convergence point, and the extracted module is where its sanitization guard belongs if this lands first. I will say so on that PR.

JiaDe-Wu:
Re-verified against current `main` (`ee7fc3bc0`), **353 commits** ahead of this branch's base. The premise is unchanged and checkable in two lines:

```python
2098: _SUMMARY_ATTEMPT_BUILDERS = {"codex_responses": _codex_summary_attempt, "anthropic_messages": _anthropic_summary_attempt}
...
2123:         build_attempt = _SUMMARY_ATTEMPT_BUILDERS.get(agent.api_mode, _chat_summary_attempt)
```

`bedrock_converse` is still not a key, so it still falls through to `_chat_summary_attempt`, whose first statement is

```python
2091:         summary_client = agent._ensure_primary_openai_client(reason=...)
```

That is the whole bug: on a Bedrock session the iteration-limit summary is built by the OpenAI client path, every time.

Measured, not asserted:

| check | result |
|---|---|
| `git cherry-pick 6281ddcfd9 c2cd769972` onto `main` | **no conflicts**, 3 files `+494/-216` |
| `pytest tests/agent/test_bedrock_iteration_limit_summary.py` on the result | **7 passed** |
| same tests with `chat_completion_helpers.py` reverted to `main` | **6 failed**, 1 passed |

One thing that run makes visible and is worth stating: without the fix, those tests emit `Failed to get summary response: Connection error` and `Auxiliary: marking nous unhealthy (payment / credit error)` — the unfixed path genuinely tries to leave the process. That is the second half of the original report (the conversation being addressed to `agent.base_url` with a Bedrock `modelId`) showing up as a test artefact. The committed fixture points at `127.0.0.1:9` precisely so the suite can't reach anything real; the noise above comes from reverting the fix underneath it, which is not a state this PR ever produces.

I'll be straight about the one thing that makes this harder to take than my other open Bedrock fixes: it is `+494/-216` and it moves 221 lines out of `chat_completion_helpers.py` into a new `chat_completion_helpers_iteration_limit.py`. The extraction is most of the diff, and it is a judgement call rather than a necessity — the fix proper is a builder function plus a dict key. If the extraction is the objection rather than the fix, say so and I'll resend it as an in-place change — no new module, no moved lines, just the builder and the key next to the two that already exist. I haven't measured that version, so I won't quote a diff size for it, but it is a small fraction of this one. I'd rather offer it than have this sit here on account of its size.

JiaDe-Wu:
Closing this in favour of **#106734**, which is the same fix and the same 7 tests without the refactor.

I said in my comment above that if the extraction was the objection rather than the fix, I'd resend it in place. Nobody asked me to — I'm doing it because on re-reading my own diff I think the trade here was wrong, and I'd rather fix that than wait for someone to spend review time telling me.

This PR is `+494/-216`. The fix proper is one builder function, one `make_client` guard, and one dict key. Everything else was moving 221 lines out of `agent/chat_completion_helpers.py` into a new `agent/chat_completion_helpers_iteration_limit.py` — a judgement call about file size, bundled into a bug fix, and it accounted for most of the 710 lines of churn. Asking a reviewer to take that in order to fix a missing dict entry is asking for the wrong trade, and in a queue this size the cost of that lands on them, not me.

#106734 is `+193/-1`: no new module, no moved lines, the two new functions sitting directly under `_chat_summary_attempt` next to the two builders that already exist. Behaviour is identical, the tests are the same ones (one import changed, one unused `import types` dropped), and it's rebased onto today's `main` rather than 353 commits behind.

Two things carried over from here that are worth keeping visible:

- The defect is confirmed still live on `main`. `_SUMMARY_ATTEMPT_BUILDERS` still names only `codex_responses` and `anthropic_messages`, so `bedrock_converse` still falls through to `_chat_summary_attempt`, whose first statement is `_ensure_primary_openai_client()`.
- The `401 Missing Authentication header` measurement in this PR's commit message — the evidence that the unpatched path really does send the conversation to `agent.base_url` — was taken on 2026-08-30 and I have not re-run it. #106734 attributes it as such rather than restating it as fresh evidence.

Nothing needed from anyone; closing it myself.

PR #98566, opened 2026-08-30
perf(state): defer sidebar preview/last_active fetch to after LIMIT in list_sessions_rich (19s -> 0.03s)

qiqikuaidianpao:
## Problem

The dashboard sidebar (`/api/sessions?order=recent`) was taking **~19 seconds** to load on a real deployment: ~1.6k listable sessions, 39k messages, 705MB `state.db` on a NAS HDD. Every page load, every time.

## Root cause

The recent-order branch of `list_sessions_rich` sorts by `COALESCE(cm.effective_last_active, s.started_at)`, which comes from the recursive compression-chain CTE. Because the ordering key is computed per candidate row, SQLite must evaluate the **whole select list for every listable session** before `ORDER BY ... LIMIT` can discard all but the page's rows — and the select list contains:

* the `_preview_raw` correlated subquery, which fetches each session's first user message and runs `SUBSTR` on its `content` (a big-payload column), and
* the `_sql_session_last_active` scalar subquery.

That's ~1.6k content-page reads per sidebar load, thrown away for all but 20 rows. Attribution by ablation on the captured SQL (removing either the chain ordering or the preview subquery individually dropped the query from 19.3s to 0.02–0.07s; the cost is the combination).

## Fix

Trim the recent-order SELECT to the compact columns + ordering key, and after `fetchall()` backfill `_preview_raw` and `last_active` for the surviving page rows with a single `WHERE s.id IN (...)` query (<= limit + backfilled pins' worth of rows). The created-order path is untouched — its `ORDER BY s.started_at` already lets LIMIT apply early, so its inline subqueries only ever ran for page rows.

## Measurements (705MB state.db, 1.6k sessions, 39k messages)

| | before | after |
|---|---|---|
| order=recent limit=20 (warm) | 19.3s | **0.03s** |
| first load after page-cache eviction | 19.3s | ~9.5s once (messages-index pages), then warm |

Correctness: same rows in the same order, same previews, same `last_active` values; on a lookup miss the backfill coalesces to `''` / `None` exactly like the old `COALESCE(..., '')` did.

## Tests

```
python -m pytest tests/test_hermes_state.py -k "last_active or list_sessions" -x -q   # 5 passed
python -m pytest tests/test_session_skill_previews.py tests/test_session_system_prompt_dedup.py -x -q  # 21 passed
```

The patch is running in production on the deployment where the 19s latency was measured (applied via a container-start patch script; this PR is the same change upstreamed).

### DIFF
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -10663,16 +10663,24 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
                 _desc = [d[0] for d in cursor.description]
                 _id_i = _desc.index("id")
                 _ids = [r[_id_i] for r in rows]
-                _ph = ",".join("?" * len(_ids))
-                _cur2 = conn.execute(
-                    "SELECT s.id AS _bid,"
-                    " COALESCE((SELECT " + _PREVIEW_RAW_SELECT + " FROM messages m"
-                    " WHERE m.session_id = s.id AND m.role = 'user' AND m.content IS NOT NULL"
-                    " AND " + _PREVIEW_ELIGIBLE_SQL +
-                    " ORDER BY m.timestamp, m.id LIMIT 1), '') AS _preview_raw,"
-                    " " + _sql_session_last_active("s") + " AS last_active"
-                    " FROM sessions s WHERE s.id IN (" + _ph + ")", _ids)
-                _bf = {r[0]: (r[1], r[2]) for r in _cur2.fetchall()}
+                # Chunk the IN(...) fetch: the dashboard router caps pages at
+                # 100, but CLI callers (sessions_cmd multiplies the requested
+                # limit by 4; console_engine passes the raw CLI limit) can
+                # exceed the classic 999 host-variable cap on SQLite builds
+                # that still default to it.
+                _bf = {}
+                for _i in range(0, len(_ids), 500):
+                    _chunk = _ids[_i:_i + 500]
+                    _ph = ",".join("?" * len(_chunk))
+                    _cur2 = conn.execute(
+                        "SELECT s.id AS _bid,"
+                        " COALESCE((SELECT " + _PREVIEW_RAW_SELECT + " FROM messages m"
+                        " WHERE m.session_id = s.id AND m.role = 'user' AND m.content IS NOT NULL"
+                        " AND " + _PREVIEW_ELIGIBLE_SQL +
+                        " ORDER BY m.timestamp, m.id LIMIT 1), '') AS _preview_raw,"
+                        " " + _sql_session_last_active("s") + " AS last_active"
+                        " FROM sessions s WHERE s.id IN (" + _ph + ")", _chunk)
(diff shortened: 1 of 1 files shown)


Enough1122:

### PR #98566 — perf(state): defer sidebar preview/last_active fetch to after LIMIT in list_sessions_rich (19s -> 0.03s)

**Summary:** Defers `_preview_raw` (first user message content page) and `last_active` subqueries to a second batched query after the `LIMIT` when `order_by_last_active` is true. The first query keeps only `COALESCE(cm.effective_last_active, s.started_at) AS _effective_last_active` for ordering, so SQLite can apply `LIMIT` without materializing the expensive select list for all ~1.6k candidates. Second query fetches previews/last_active for only the page's ids via `WHERE s.id IN (...)`. Created-order path (`ORDER BY s.started_at`) is untouched where `LIMIT` already applies early.

- Order is preserved: `_bf` is a dict keyed by id, rows are rebuilt via `dict(zip(_desc + [...], tuple(r) + _bf.get(...)))` in original `rows` order, so page order from the `ORDER BY effective_last_active` survives the `IN` fetch (which is unordered) (`hermes_state.py:48-53`).
- `IN` placeholder string `",".join("?" * len(_ids))` assumes `len(_ids) <= SQLITE_MAX_VARIABLE_NUMBER` (999). The page's `LIMIT` is small in dashboard usage (`/api/sessions?order=recent`), so safe; if a caller ever raises the limit beyond ~900 the second query would hit `too many SQL variables`. Consider chunking or asserting `len(_ids) < 900` for defensive hardening — low risk today.
- Guard `if order_by_last_active and rows:` correctly skips the second roundtrip on empty result and on the `s.started_at` path. `cursor.description` is still valid after `fetchall()` — stable.
- `_PREVIEW_RAW_SELECT` / `_PREVIEW_ELIGIBLE_SQL` and `_sql_session_last_active("s")` are reused verbatim, so eligibility and preview semantics are unchanged. Missing ids fallback to `("", None)` — matches the original `COALESCE(..., '')` for preview and `NULL` for last_active.
- Uses the same `_read_ctx` connection for both queries, so the snapshot is consistent within the read transaction. No write lock taken.

**Verdict:** Correct perf fix with large measured win (19s→30ms warm on 705 MB NAS DB). No blocking issues.

qiqikuaidianpao:
Thanks for the review — both observations addressed:

- **Order preservation**: agreed, and unchanged by the new commit — `_bf` is only consulted by key when rebuilding each row in the original `rows` order.
- **Host-variable cap**: good catch on the latent bound. The dashboard router caps pages at 100 ids, but two CLI callers don't: `sessions_cmd` fetches `max(limit * 4, limit)` and `console_engine` forwards the raw CLI limit, so a large page on a SQLite build still defaulting to `SQLITE_MAX_VARIABLE_NUMBER=999` would have failed. 6c1123f chunks the id list in groups of 500 rather than asserting, so large pages keep working on every build.

All 26 previously-passing tests still pass (5 recent-order + 21 preview/dedup).

PR #98527, opened 2026-08-30
fix(memory): align Hindsight client pin at 0.9.2

ttomiczek:
## Summary

Updates Hermes's Hindsight client from the stale `0.6.1` pin to the reviewed `0.9.2` release across every enforcement path:

- the `hindsight` optional dependency and committed `uv.lock`
- the Hindsight provider minimum and plugin manifest
- `memory.hindsight` in `tools/lazy_deps.py`
- the relative `exclude-newer` exemption required to resolve this exact reviewed release immediately

Adds a regression test covering the pin, lock resolution, and `exclude-newer` exemption.

This prevents normal updates and lazy dependency healing from restoring the obsolete 0.6.1 client after an operator has moved to the 0.9.x Hindsight stack.

## Verification

```text
92 passed in 5.53s
uv lock --check
```

Refs #86992

Issue #95855, opened 2026-08-26
fastmcp needs mcp 1.x but pyproject pins mcp==2.0.0 -> Hindsight local_embedded broken after every update (request_ctx missing)


LauraAssistant:
## Summary
After `hermes update`, the Hindsight `local_embedded` memory provider shows `Status: not available ✗` in `hermes memory status` and "hindsight configured but not available" in `hermes doctor`. The root cause is an internally inconsistent dependency pin, not a broken install.

## Environment
- Hermes v0.20.5 (2026.8.19), git install, Windows 11, Python 3.11.15 venv

## Symptom
```
$ hermes memory status
  Plugin:    installed ✓
  Status:    not available ✗
  Missing:
    ✗ HINDSIGHT_API_KEY  → https://ui.hindsight.vectorize.io   <- red herring, see below
```

Real error chain (visible when probing the venv):
```
import hindsight
  -> fastmcp/__init__.py __getattr__: raise ImportError("FastMCP server support is not installed...")
     from exc: fastmcp/server/context.py, line 15:
       from mcp.server.lowlevel.server import request_ctx
       ImportError: cannot import name 'request_ctx' from 'mcp.server.lowlevel.server'
```

`is_available()` -> `_check_local_runtime()` -> `import hindsight` fails -> provider marked unavailable, recall pipeline disabled.

## Root cause: version pin conflict
`pyproject.toml` pins `mcp==2.0.0` (in the `mcp` and `computer-use` extras), but:
1. `fastmcp` 3.4.5 / 3.4.7 imports `request_ctx` from `mcp.server.lowlevel.server` — **mcp 2.0.0 removed this symbol entirely** (grep across the whole installed mcp 2.0.0 package: zero occurrences). It exists in mcp 1.x.
2. `claude-agent-sdk` (fastmcp's dependency) requires `mcp<2.0.0,>=1.23.0` — `pip check` confirms the conflict.

So the pinned dependency graph can never import cleanly: fastmcp needs mcp 1.x, Hermes pins mcp 2.0.0.

## Extra notes
- The embedded daemon itself still runs: the manager falls back to `uvx hindsight-api@<version>`, whose isolated env resolves the consistent set `mcp 1.29.1 + fastmcp 3.4.7`. But the in-venv `_check_local_runtime()` probe still fails, so the plugin reports unavailable regardless.
[...]

PR #98569, opened 2026-08-30
fix(hindsight): preserve the official embedded lifecycle with MCP 2

neilinger:
## Summary

Fixes the supported Hindsight `local_embedded` lifecycle without putting Hindsight's server dependency graph into Hermes' MCP 2 environment.

- install `hindsight-client==0.9.2` for all Hindsight modes
- install `hindsight-embed==0.9.2` for `local_embedded`
- use Hindsight's upstream embed manager to launch the matching `uvx hindsight-api@0.9.2` server out of process
- stop installing `hindsight-all` into Hermes' managed environment
- give provider-aware setup/update installs a 600-second timeout
- align plugin metadata, lazy dependency registration, extras, lockfile, docs, setup UI, and updater behavior
- preserve the `HindsightEmbedded.client` wrapper contract in the lightweight adapter

## Why

Hermes currently pins MCP 2, while the `hindsight-all` server stack resolves through MCP 1.x. Installing both into one environment either downgrades/breaks Hermes' MCP surface or fails dependency checks. Hindsight's published `hindsight-embed` manager already provides the intended boundary and starts a version-matched API through `uvx`, so Hermes can retain lifecycle ownership without maintaining a private server integration.

Related: #95855. This also subsumes the dependency-version part of #98527.

## Verification

- 109 targeted setup/provider/packaging tests passed
- 167 memory-setup web-server tests passed
- follow-up provider regression suite: 85 passed
- Ruff passed for changed Python files
- real disposable `hermes memory setup hindsight` selected `local_embedded` and exited 0
- installed environment verified as:
  - `mcp==2.0.0`
  - `hindsight-client==0.9.2`
  - `hindsight-embed==0.9.2`
  - no `hindsight-all`, `hindsight-api-slim`, or `fastmcp-slim` in Hermes
- `uv pip check`: compatible
- provider resolved server command `uvx hindsight-api@0.9.2`
- `/health`: HTTP 200 with database connected
- forced update-time dependency refresh re-established the same package set

Eight broad update-command tests also fail unchanged on clean `origin/main` because they inspect the live macOS launchd gateway instead of an isolated fixture; they are not introduced by this patch.

### DIFF
--- a/hermes_cli/memory_setup.py
+++ b/hermes_cli/memory_setup.py
@@ -17,6 +17,9 @@ from hermes_constants import get_hermes_home
 from hermes_cli.secret_prompt import masked_secret_prompt
 
 _CANCELLED = -1
+_DEFAULT_DEPENDENCY_INSTALL_TIMEOUT = 120
+_HEAVY_DEPENDENCY_INSTALL_TIMEOUT = 600
+_HINDSIGHT_EMBED_SPEC = "hindsight-embed==0.9.2"
 
 
 def _provider_pip_dependencies(provider_name: str, declared: list) -> list:
@@ -24,11 +27,11 @@ def _provider_pip_dependencies(provider_name: str, declared: list) -> list:
 
     ``plugin.yaml`` declares the provider's baseline bridge packages, but
     some providers install mode-dependent extras at setup time that the
-    manifest can't express. Hindsight's ``local_embedded`` mode installs
-    ``hindsight-all`` (daemon + embedder + client) during
-    ``hermes memory setup`` — if the update-time refresh only reinstalled
-    the declared ``hindsight-client``, the embedded daemon would stay
-    broken after a venv rebuild stripped ``hindsight-embed`` (#70636).
+    manifest can't express. Hindsight's ``local_embedded`` mode needs its
+    lightweight embed manager in addition to the client. The manager launches
+    ``hindsight-api`` out of process (using its version-matched ``uvx`` fallback),
+    which keeps the API's MCP-1 dependency tree separate from Hermes' MCP-2
+    client while preserving the official setup/update lifecycle (#95855).
     """
     deps = list(declared or [])
     if provider_name == "hindsight":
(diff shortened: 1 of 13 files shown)


teknium1:
Closing: the bundled Hindsight provider has moved out of this repo, and the core-side part of this PR is already on `main` or no longer has anything to attach to.

Thanks @neilinger for this contribution. In #119888 (merge `9d799e0531c`; removal commit `4cbf862abe4`) the in-tree `plugins/memory/hindsight/` provider was removed — Hindsight now installs from the plugin catalog and its code lives in [vectorize-io/hindsight `hindsight-integrations/hermes`](https://github.com/vectorize-io/hindsight/tree/dc75038866a3966b5bf5c82ff57c7a758bb4070a/hindsight-integrations/hermes) (maintained by @nicoloboschi). There is no longer any code in this repo for the Hindsight half of this PR to patch, so we are closing every open PR against the bundled provider rather than leaving them stranded.

Triage notes:
- Replaces hindsight-all with hindsight-client==0.9.2 + hindsight-embed==0.9.2 (out-of-process uvx daemon via a _ManagedEmbeddedHindsight adapter) and bumps pins in memory_setup/update_cmd/web_server/lazy_deps/pyproject/uv.lock. Every core hunk existed only to serve the bundled provider and its target code is gone on origin/main (removed in 4cbf862abe4; only the generic pip→import name map remains in memory_setup.py:20), so nothing survives to merge here. The pinned upstream tree still installs hindsight-all (embedded.py:133, setup.py:118), so the MCP-conflict fix and the MCP-2-safe package split are worth proposing to vectorize-io/hindsight.
- Core half: origin/main hermes_cli/memory_setup.py::_IMPORT_NAMES (line 20) is the only hindsight reference left — the mode-dependent hindsight-all branch in _provider_pip_dependencies and the 120s/600s timeout helper have no host; tools/lazy_deps.py has no "memory.hindsight" pin; hermes_cli/web_server.py and pyproject.toml have zero hindsight hits; all removed in 4cbf862abe4 (chore(memory): remove the bundled hindsight provider) / #119888 9d799e0531c.
- **Still relevant at the catalog pin (`dc750388`)?** Yes — the same code is at `hindsight-integrations/hermes/embedded.py:133` in the upstream tree. It is listed with your credit in https://github.com/vectorize-io/hindsight/issues/4662 so it is not lost; if you want to carry the fix yourself, please open it against vectorize-io/hindsight — it would be welcome there.

If you believe this was closed in error, comment and we will reopen.

(Bulk-closed in the hindsight-move close pass.)

PR #94072, opened 2026-08-24
fix(models): honor excluded_providers for the virtual MoA picker row

liuhao1024:
## What does this PR do?

`list_picker_providers()` applies `model_catalog.excluded_providers` inside `list_authenticated_providers()` and then prepends the virtual MoA row afterwards — so the gateway `/model` picker (Telegram/Discord), which passes `include_moa=True` unconditionally, always showed the "Mixture of Agents" button even when the user excluded `moa`. The CLI `hermes model` picker already honors the exclusion; only the gateway path bypassed it, because the virtual MoA row is the one provider row constructed outside the exclusion-filtered path.

This guards the prepend with the same normalized membership check `list_authenticated_providers()` uses for its exclusion set (`str(p).strip().lower()`), so `excluded_providers: ["moa"]` (or a case/whitespace variant) now hides the row on gateway pickers too. No gateway-side change is needed — `gateway/slash_commands.py` already forwards `excluded_providers`.

## Related Issue

Fixes #94068

## Type of Change

- [x] 🐛 Bug fix

## Changes Made

- `hermes_cli/model_switch.py` (`list_picker_providers`) — the MoA prepend is now also gated on `"moa"` not appearing in the normalized `excluded_providers` set, with a comment naming the bypass (#94068). Normalization matches the existing set construction in `list_authenticated_providers`.
- `tests/hermes_cli/test_list_picker_providers.py` — new tests for the virtual-row exclusion:
  - `test_moa_row_hidden_when_excluded` — the reported gateway scenario (`include_moa=True` + `excluded_providers=["moa"]` → no MoA row).
  - `test_moa_row_hidden_with_case_and_whitespace_variant` — pins the normalization contract.
  - `test_moa_row_present_without_exclusion` — default prepend behaviour unchanged (row first).
  - `test_moa_row_survives_other_provider_exclusion` — excluding an unrelated provider does not hide the MoA row.

## How to Test

- Run: `.venv/bin/python -m pytest tests/hermes_cli/test_list_picker_providers.py -q`
- Observed result: `8 passed`. Fail-on-main check (`git stash` of the source change): the two exclusion tests fail against unpatched `main`; the two behaviour-preservation tests pass both ways.
- Regression sweep (`test_list_picker_providers.py`, `test_model_catalog.py`, `test_model_switch_custom_providers.py`, `tests/gateway/test_model_picker_persist.py`, `tests/gateway/test_25107_stale_base_url_api_mode.py`): `100 passed` total.

## Checklist

[...]

PR #98572, opened 2026-08-30
fix(model-picker): honor excluded_providers for the virtual MoA row

darblex:
## Bug

`model_catalog.excluded_providers` filters real providers out of the model picker (`hermes model`), but the synthetic `moa` (Mixture-of-Agents) row is injected *after* that filtering pass in `list_picker_providers()`, so adding `moa` to `excluded_providers` had no effect — the MoA row still showed up.

## Fix

Compute the excluded-provider set once, before the MoA injection branch, and skip injecting the row when `moa` is in it — the same treatment real providers already get.

## Testing

- Added `test_excluded_providers_hides_virtual_moa_row` (regression test).
- Full `tests/hermes_cli/test_model_switch_custom_providers.py` suite: 71 passed, 0 failed, on top of current `main` (66666f6e2e).

Small, targeted, single call-site fix — no other behavior changed.

### DIFF
--- a/hermes_cli/model_switch.py
+++ b/hermes_cli/model_switch.py
@@ -4008,7 +4008,12 @@ def list_picker_providers(
         for_picker=True,
         excluded_providers=excluded_providers,
     )
-    if include_moa:
+    excluded = {
+        str(provider).strip().lower()
+        for provider in (excluded_providers or [])
+        if str(provider).strip()
+    }
+    if include_moa and "moa" not in excluded:
         providers = _prepend_moa_picker_provider(providers, current_provider=current_provider)
 
     filtered: List[dict] = []
--- a/tests/hermes_cli/test_model_switch_custom_providers.py
+++ b/tests/hermes_cli/test_model_switch_custom_providers.py
@@ -19,6 +19,7 @@ from hermes_cli.model_switch import (
     _fetch_picker_live_models,
     _save_discovered_models_to_config,
     list_authenticated_providers,
+    list_picker_providers,
     switch_model,
 )
 from hermes_cli.providers import resolve_provider_full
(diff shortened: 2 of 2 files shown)


Enough1122:

### PR #98572 — fix(model-picker): honor excluded_providers for the virtual MoA row

**Summary:** `list_picker_providers(excluded_providers=...)` filtered real providers via `model_catalog` but injected the synthetic `moa` (Mixture-of-Agents) row afterwards in `_prepend_moa_picker_provider`, so `excluded_providers=["moa"]` had no effect. Now computes a normalized `excluded` set (`str(...).strip().lower()` with empty-string filter) up front and gates the injection with `if include_moa and "moa" not in excluded` (`hermes_cli/model_switch.py:10-15`). Test pins the fix.

- Normalization matches the slug `"moa"` lowercase check; real provider slugs are already lowercased by the catalog, so consistent. `str(provider).strip()` guard (`model_switch.py:12-13`) tolerates `None`/non-string entries without raising — aligns with `excluded_providers` coming from config where entries may be mixed.
- `excluded_providers or []` (`model_switch.py:12`) handles `None` correctly, mirroring the catalog's own handling. Duplicates and whitespace-only entries are collapsed by the set comprehension.
- `current_provider` is still forwarded to `_prepend_moa_picker_provider` so MoA highlighting is preserved when not excluded. The `filtered` list comprehension below is unchanged — MoA exclusion is applied before filtering/sorting, so no second filtering pass is needed.
- Minimal diff (`+5 -1` logic, `+6` test); risk is near zero. The test uses `list_authenticated_providers` mock returning `[]` to isolate the virtual row — appropriate.

**Verdict:** Minimal, correct fix for the documented `excluded_providers` bypass on the virtual row. No blocking issues.

alt-glitch:

Duplicate of #94072 — both patches apply the same excluded_providers membership check before injecting the virtual MoA picker row. #94072 is the earlier open PR and centralizes the normalization helper.

PR #33729, opened 2026-05-28
fix(error_classifier): fast-fail 503/529 capacity errors to fallback chain

OmarB97:
## Why

When Nous Portal (or any upstream provider) returns HTTP 503 with an 'upstream capacity limits' message, the error classifier marks it `retryable=True`. This causes the conversation loop to retry with exponential backoff for 2+ minutes before falling back to the next configured provider. Every retry hits the same capacity wall — wasted time and API quota.

Real-world reproduction: user on `qwen3.7-max` via Nous gets 503 → Hermes burns 2 minutes retrying → eventually falls back to kimi-k2.6 on opencode-zen. With this fix, the fallback happens in under 1 second.

### `agent/error_classifier.py`

When classifying 503/529 errors, the classifier now checks for capacity-pattern keywords (`upstream capacity`, `capacity limits`) in the error message. If matched, the error is classified as `retryable=False, should_fallback=True` instead of the blanket `retryable=True` that triggers exponential backoff.

### `agent/conversation_loop.py` (lines 2550-2563)

Added an eager-fallback check for overload errors (mirroring the existing rate-limit eager-fallback at lines 2530-2547). When `FailoverReason.overloaded` is classified as non-retryable, the conversation loop immediately falls back to the next provider in the configured fallback chain instead of entering the retry-with-backoff path.

### `tests/agent/test_error_classifier.py`

3 new tests:
- `test_503_upstream_capacity_fast_fail`: 503 + capacity pattern → retryable=False, should_fallback=True
- `test_529_upstream_capacity_fast_fail`: 529 + capacity pattern → retryable=False, should_fallback=True
- `test_503_without_capacity_pattern_still_retries`: 503 without capacity pattern → retryable=True (preserves existing behavior for transient non-capacity 503s)

## How to review

The classifier change is 6 lines (pattern match + early return). The conversation loop change is a 12-line block inserted between the existing rate-limit eager fallback and the general retry logic. Both follow the established patterns in their respective files.

## Evidence

- 153 error classifier tests pass (150 existing + 3 new)
- Observed in production: Nous 503 'upstream capacity limits' → kimi-k2.6 fallback in <1s after patch vs 2+ minutes before

## Risks

Low. The change only affects 503/529 errors that explicitly contain capacity-pattern keywords. All other 503/529 errors (transient, non-capacity) retain the existing `retryable=True` behavior.

PR #98575, opened 2026-08-30
fix(agent): fail over cron turns on provider timeouts

albidev:
## Problem

Cron jobs with a configured provider fallback chain never fail over when the
primary provider times out. The turn exhausts its retries on the dead
provider and dies with:

```
RuntimeError: Non-streaming API call timed out after 180s with no response (threshold: 180s)
API calls: 0
```

No fallback provider is ever attempted, even though `fallback_providers` is
configured. A nightly job observed in production: primary Ollama Cloud
timed out after 180s, `API calls: 0`, duration 549s, and the configured
fallback chain (Nous Portal → OpenRouter → local oMLX) was never reached.

## Root cause

In `agent/error_classifier.py`, transport/timeout errors are classified as
`retryable=True` but `should_fallback=False`:

- `_TRANSPORT_ERROR_TYPES` and `isinstance(error, (TimeoutError, ConnectionError, OSError))`
- `_TIMEOUT_MESSAGE_PATTERNS`
- `_CONNECTION_MESSAGE_PATTERNS`

A retryable-but-no-fallback classification makes the conversation loop keep
retrying the same failing provider until the turn budget is exhausted, then
ends the turn — the fallback chain is never activated. Every other
recoverable error class (429, 5xx) already requests fallback; timeouts are
the only path that gets stuck retrying into the void.

## Fix

Add `should_fallback=True` to all three timeout/transport/connection
classification paths. A provider that cannot respond within the deadline
is no more likely to respond on the next attempt than one returning 429;
it should hand off to the next configured provider instead of burning the
turn.

## Tests

Added regression test `test_timeout_is_retryable_and_requests_provider_fallback`
asserting timeouts remain retryable but now request provider fallback.

- `pytest tests/agent/test_error_classifier.py` — 123 passed
- `pytest tests/run_agent/test_provider_fallback.py` — 46 passed
- `pytest tests/run_agent/test_32646_fallback_429_after_timeout.py` — 5 passed

## Related

- #33729 — fast-fail 503/529 capacity errors to fallback chain (complementary:
  that PR covers server errors, this covers transport timeouts)

### DIFF
--- a/agent/error_classifier.py
+++ b/agent/error_classifier.py
@@ -1231,7 +1231,11 @@ def classify_api_error(
     # ── 8. Transport / timeout heuristics ───────────────────────────
 
     if error_type in _TRANSPORT_ERROR_TYPES or isinstance(error, (TimeoutError, ConnectionError, OSError)):
-        return _result(FailoverReason.timeout, retryable=True)
+        return _result(
+            FailoverReason.timeout,
+            retryable=True,
+            should_fallback=True,
+        )
 
     # ── 9. Fallback: unknown ────────────────────────────────────────
 
@@ -2048,7 +2052,11 @@ def _classify_by_message(
     # loop rebuilds the client instead of treating the turn as an empty
     # model response.
     if any(p in error_msg for p in _TIMEOUT_MESSAGE_PATTERNS):
-        return result_fn(FailoverReason.timeout, retryable=True)
+        return result_fn(
+            FailoverReason.timeout,
+            retryable=True,
+            should_fallback=True,
+        )
 
     # Connection-establishment / DNS failure message patterns — same shim
     # problem as the timeout patterns above: the wrapping exception type is
@@ -2058,7 +2066,11 @@ def _classify_by_message(
     # client rebuild apply. Never routes to compression: a connection that
     # was never established is not a context-overflow signal.
     if any(p in error_msg for p in _CONNECTION_MESSAGE_PATTERNS):
-        return result_fn(FailoverReason.timeout, retryable=True)
+        return result_fn(
+            FailoverReason.timeout,
+            retryable=True,
+            should_fallback=True,
+        )
 
     return None
 
--- a/tests/agent/test_error_classifier.py
+++ b/tests/agent/test_error_classifier.py
@@ -459,6 +459,16 @@ class TestClassifyApiError:
         assert result.reason == FailoverReason.timeout
         assert result.retryable is True
 
+    def test_timeout_is_retryable_and_requests_provider_fallback(self):
+        """A stale non-streaming provider call must hand off to the next
+        provider instead of exhausting the cron turn on the same backend."""
+        result = classify_api_error(
+            TimeoutError("Non-streaming API call timed out after 180s with no response")
+        )
+        assert result.reason == FailoverReason.timeout
+        assert result.retryable is True
+        assert result.should_fallback is True
+
     def test_400_bad_request_still_non_retryable_format_error(self):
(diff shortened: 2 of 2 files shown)


Enough1122:

### PR #98575 — fix(agent): fail over cron turns on provider timeouts

**Summary:** Cron turns with `fallback_providers` exhausted retries on a dead primary and died with `RuntimeError: Non-streaming API call timed out after 180s … API calls: 0` without ever trying the next provider. Timeout-family errors were classified as `retryable=True` but `should_fallback=False`, so the turn-level retry loop rebuilt the client instead of handing off. This PR sets `should_fallback=True` on the three timeout branches in `agent/error_classifier.py`: the transport/type branch (`TimeoutError`/`ConnectionError`/`OSError` or `_TRANSPORT_ERROR_TYPES`, `:1231`) and the two message-pattern shims (`_TIMEOUT_MESSAGE_PATTERNS` and `_CONNECTION_MESSAGE_PATTERNS` in `_classify_by_message`, `:2051-2042`).

- The three sites now return `_result(FailoverReason.timeout, retryable=True, should_fallback=True)` (`error_classifier.py:9-14,22-27,35-40`). This keeps the existing `retryable` semantics (so non-cron turns can still retry) while enabling the fallback-provider chain for timeouts — the stated intent ("a stale non-streaming provider call must hand off", `test_error_classifier.py:52`). Connection-establishment/DNS failures via `_CONNECTION_MESSAGE_PATTERNS` are correctly grouped with timeouts and also excluded from compression routing (comment `:2030`).
- Scope is intentionally broad: all timeout-like errors now request fallback, not just `Non-streaming API call timed out`. This matches the PR description's "provider fallback chain never fail over when the primary provider times out" — interactive turns will also fall back on a 180s stall rather than exhausting retries on the same backend. If that is too aggressive for interactive use, a cron-vs-interactive distinction could be introduced later, but the current change matches the fallback-provider contract (fallback is opt-in via config).
- `OSError` is included (`error_classifier.py:8`), which also covers `BrokenPipeError`/`ConnectionResetError` — reasonable as transport failures. No change to 400/401/403/etc branches; their `should_fallback=False` remains correctly guarded (`test_400_bad_request_still_non_retryable` stays green).
- Test `test_timeout_is_retryable_and_requests_provider_fallback` (`test_error_classifier.py:459`) pins the new `should_fallback` flag on the exact production message `"Non-streaming API call timed out after 180s with no response"` (`:56`), tying the fix to the observed nightly Ollama Cloud outage.

**Verdict:** Targeted error-classifier fix for the observed cron fallback gap. No blocking issues.

kshitijk4poor:
Closing as **not planned** — the change is a no-op on the code path it targets.

The PR adds `should_fallback=True` to three timeout/transport branches in `agent/error_classifier.py`, but `classified.should_fallback` is never consumed by the conversation loop's fallback decision. The only read is a `logger.debug()` call. The actual decision at `agent/conversation_loop.py` (~L5576) computes its own predicate:

```python
_is_transport_failure = classified.reason in {FailoverReason.timeout, FailoverReason.overloaded}
_should_fallback = (is_rate_limited and ...) or (_is_transport_failure and retry_count >= 2)
```

So provider timeouts on cron turns **already fail over** after 2 loop retries via `FailoverReason.timeout` — the stated goal of this PR is existing behavior, and the new field doesn't change any decision. (The cron inactivity watchdog, `HERMES_CRON_TIMEOUT`, is a separate mechanism and also unaffected.)

Verified against `origin/main@82c95a2a01`. Thanks @albidev — if you've observed cron turns *not* failing over on timeout in practice, that would be a valuable bug report with logs (the failure would be in the retry accounting, not the classifier).

PR #98576, opened 2026-08-30
fix(streaming): read reasoning from delta.model_extra for custom OpenAI-compatible providers

### LINKED ISSUE #98224: [Bug] Custom providers drop reasoning in streaming — delta has no reasoning field after first tool call
## Bug Description

Reasoning models routed through **custom OpenAI-compatible providers** silently lose their reasoning stream in the Desktop GUI — the thinking block is missing or mixed into body text after the first tool call. The same models work correctly through the built-in DeepSeek provider.

**The bug is not model-specific and does not depend on any user-provided plugin file**: it reproduces with the official bundled kernel + any custom provider configured in `config.yaml` (`providers.<name>` block) that fronts a reasoning model. (I stripped any custom user plugin I had before testing; the official bundled provider profile reproduces it too.)

## Steps to Reproduce

1. Add a custom provider to `config.yaml`, e.g.:

```yaml
providers:
  mygateway:
    api: https://<gateway-host>/provider/v1
    key_env: MYGATEWAY_API_KEY
    transport: chat_completions
    default_model: deepseek/deepseek-v4-flash
```

2. Start a conversation in the Desktop GUI with that provider + a reasoning model (`deepseek-v4-flash` / `deepseek-v4-flash-vision-exp`), set a high reasoning effort.
3. Send a message that requires multiple tool calls (think → tool call → think again → output).
4. Observe the GUI:
   - The **first** thinking block appears correctly separated.
   - After the **first tool call**, subsequent thinking content is rendered **as regular body text** (styled identically to the answer) instead of inside a thinking block — or the thinking block vanishes. `reasoning_content` is empty in the DB for those turns.

## Expected Behavior

Thinking content is always rendered inside its dedicated thinking-block style, separated from body text, on every turn including after tool calls — like the built-in DeepSeek provider does (`api.deepseek.com`, `reasoning_content` deltas, works).

## Actual Behavior

After the first tool call, thinking text is rendered as body text, or the thinking block disappears entirely.

## Affected Component

Agent Core (conversation loop / streaming); Desktop GUI rendering; custom provider (OpenAI-compatible gateway) stream handling.

## Root Cause Investigation (runtime probe, official bundled kernel)

I instrumented `agent/chat_completion_helpers.py` around the streaming delta accumulation (`PROBE2` log). Streaming `deepseek-v4-flash-vision-exp` via a custom gateway (OpenAI-compatible `/provider/v1`):

```
PROBE2 chunk_is_dict=False delta_is_dict=False
  chunk_keys=['id', 'choices', 'created', 'model', 'object', 'service_tier', 'system_fingerprint', 'usage']
  delta_keys=['content', 'function_call', 'refusal', 'role', 'tool_calls']
  has_reasoning_attr=False has_reasoning_content_attr=False
```

Key observations:

- The SDK-parsed `delta` object (OpenAI SDK 2.24.0) has **no `reasoning` / `reasoning_content` field at all** through the custom gateway, even though a **bare `OpenAI(...)` client** streaming the same endpoint **does** parse `delta.reasoning` (e.g. 648 `delta.reasoning` hits; `model_extra` contains `reasoning`/`reasoning_details`).
- This reproduces with the **official bundled** provider profile for the same gateway (no user-added plugin) — so a user plugin is NOT required.
- The built-in **DeepSeek provider** (`api.deepseek.com`, `deepseek-v4-flash-vision-exp`) returns `reasoning_content` deltas and renders correctly.
- The same symptom also appears via another custom gateway (OpenAI-compatible, fronting `zai/glm-5.3-flash`) — `reasoning_tokens=0`, no thinking block. Consistent with a provider/stream handling difference, not model-specific.

So the loss happens when Hermes steps the stream for a **custom provider**: the reasoning delta reaches the endpoint's response but does not survive into the SDK-parsed `delta` object that `chat_completion_helpers.py` reads (`getattr(delta, "reasoning_content", None) or getattr(delta, "reasoning", None)` — both return None).

## Debug Report (local summary — I did not upload to a public paste service)

```
os:               Windows 10 AMD64
python:           3.11.15
openai_sdk:       2.24.0
profile:          default
model:            deepseek/deepseek-v4-flash-vision-exp
provider:         commandcode
terminal:         local
gateway:          running
```

## Environment / Timing

- Hermes Agent v0.20.6 (git, upstream 4209d371); Windows 10; Python 3.11.15; OpenAI SDK 2.24.0; Desktop GUI.
- Custom providers affected: an OpenAI-compatible gateway (`https://api.commandcode.ai/provider/v1`, fronting `deepseek/deepseek-v4-flash-vision-exp`); another (`https://api-gateway.merge.dev/v1`, fronting `zai/glm-5.3-flash`). Both reproduce. Built-in DeepSeek provider: works.
- Affected sessions (UTC): 2026-08-29 13:54 → 2026-08-29 21:52 UTC (multiple sessions, both gateways; the symptom appears consistently across them).

## Notes for maintainers

- The discrepancy (a bare client parses `delta.reasoning`; Hermes' client does not) suggests Hermes' client construction / chunk handling differs from a stock OpenAI client for custom providers (possibly request parameters, headers, or how the delta is dereferenced in `chat_completion_helpers.py`).
- A provider-level `build_api_kwargs_extras` forwarding `reasoning_effort`/`thinking` did NOT change the symptom — the effort forwarding is not the culprit; the stream delta extraction is.
- Custom gateways return reasoning under the non-standard `reasoning` field name, whereas the built-in DeepSeek endpoint returns `reasoning_content`. Hermes' streaming path appears to lose the former.

salch-cred:
## Problem (#98224)

Custom OpenAI-compatible gateways that front reasoning models (DeepSeek, GLM) forward reasoning as an extra SSE field. The OpenAI SDK promotes it from \delta.model_extra\ to a named attribute **only when it recognises the provider** in its schema. For unknown/custom providers, reasoning lands only in \model_extra['reasoning']\ — NOT on \delta.reasoning\ or \delta.reasoning_content\.

The streaming accumulator in \gent/chat_completion_helpers.py\ read only the named attributes, so **streaming reasoning was silently dropped after the first tool call** on custom providers. The built-in \pi.deepseek.com\ provider worked because the SDK promotes the field for known endpoints.

## Fix

Add a \model_extra\ fallback in the streaming delta loop (lines 4251–4264). When neither \delta.reasoning_content\ nor \delta.reasoning\ is set, check \delta.model_extra\ for the same keys. No behaviour change for built-in providers where the SDK already promotes the field.

Fixes #98224

### DIFF
--- a/agent/chat_completion_helpers.py
+++ b/agent/chat_completion_helpers.py
@@ -4248,8 +4248,27 @@ def interruptible_streaming_api_call(agent, api_kwargs: dict, *, on_first_delta=
             if hasattr(chunk, "model") and chunk.model:
                 model_name = chunk.model
 
-            # Accumulate reasoning content
-            reasoning_text = getattr(delta, "reasoning_content", None) or getattr(delta, "reasoning", None)
+            # Accumulate reasoning content.
+            # Primary: SDK-promoted attributes (native DeepSeek / Moonshot).
+            # Fallback: model_extra dict — custom OpenAI-compatible gateways
+            # (e.g. a relay fronting deepseek-v4-flash) forward reasoning as
+            # an extra SSE field.  The SDK (openai>=1.54) promotes known extras
+            # from its known-model schema; an *unknown* provider is not in the
+            # schema, so reasoning lands in model_extra but NOT on the delta
+            # object as a named attr.  Without this fallback, streaming
+            # reasoning is silently dropped after the first tool call on custom
+            # providers (#98224).
+            reasoning_text = (
+                getattr(delta, "reasoning_content", None)
+                or getattr(delta, "reasoning", None)
+            )
+            if not reasoning_text and hasattr(delta, "model_extra"):
+                _delta_extra = getattr(delta, "model_extra", None) or {}
+                if isinstance(_delta_extra, dict):
+                    reasoning_text = (
+                        _delta_extra.get("reasoning_content")
+                        or _delta_extra.get("reasoning")
+                    )
             if reasoning_text:
                 # Summary-part models (gpt-5.x and other Responses relays) send
                 # one complete markdown block per delta with no separator, so


Enough1122:

**Overall:** Focused fix for custom OpenAI-compatible relays dropping streaming `reasoning_content` after tool calls (#98224). `model_extra` fallback is the right layer given `openai>=1.54` only promotes known schema fields.

**Correctness:**
- `agent/chat_completion_helpers.py:25-31` — fallback gated by `hasattr(delta, "model_extra")` and `isinstance(..., dict)` avoids AttributeError on SDK variants without `model_extra`. Order `reasoning_content` → `reasoning` matches primary path.
- Uses `or` chaining consistently with primary path, so empty-string still falls through.

**Non-blocking nit:**
- Consider guarding against `model_extra` being a pydantic `Extra` object rather than dict on future SDK bumps — current `isinstance(dict)` check already fails closed, which is safe but would silently skip if SDK changes representation. No action needed now.

**Verdict:** LGTM — small, well-commented, no behavior change for native SDK attributes.

kshitijk4poor:
Thanks @salch-cred — current `main` already fixes this via 36c987ea6e ("fix: reasoning-only streams whose deltas carry reasoning only in model_extra no longer fail as empty"), so this PR is superseded and I'm closing it. Your PR predates that commit, so credit to you for pinning it first.

PR #98577, opened 2026-08-30
fix(desktop): clear queue-drain failure counters on mount to prevent misleading banner after restart

### LINKED ISSUE #98015: Desktop: queued-prompt drain after restart fails against reaped session runtime, then shows misleading 'message not sent' banner
### Summary

On the macOS desktop app (v0.20.6), if a prompt sits in the composer queue when the app exits, restarting the app makes the background queue-drain submit the prompt to a session whose runtime no longer exists (backend was restarted alongside the app, so every in-memory session runtime is gone). The submit is rejected, the drain retries up to `MAX_AUTO_DRAIN_ATTEMPTS` times, then shows the `queueStuckTitle` / `queueStuckBody` banner:

> 排队消息未发送 — 排队的对话多次发送失败。它仍在队列中，请重试发送。

The message is **not** lost and the queue is **not** broken — the user just sees a scary error banner on every login after having left an unsent draft.

### Root cause

- The composer queue is persisted in `localStorage` (`hermes.desktop.composerQueue.v1` in `apps/desktop/src/store/composer-queue.ts`), so it survives restarts.
- Session runtimes are in-memory only and are torn down when the desktop app restarts its backend (`python -m hermes_cli.main --profile <p> serve ...`). The gateway logs each rejected call as:

  ```
  WARNING tui_gateway.server: session-scoped RPC rejected: method=process.list/slash.exec/approval.pending session_id='<id>' not in memory (detached/reaped runtime; client should resume the stored session)
  ```

- `use-background-queue-drain.ts` calls `submitText(..., { sessionId: runtimeSessionId, storedSessionId })` where `runtimeSessionId` is null after restart; the backend rejects instead of resuming the stored session, `accepted === false`, and after 4 failures `notify({ kind: 'error', title: t.composer.queueStuckTitle, ... })` fires (see `onFail` in `apps/desktop/src/app/session/hooks/use-background-queue-drain.ts`).

### Repro

1. Type a message in a desktop session, send it, and quit the app **before the turn finishes** (or have any prompt queued).
2. Relaunch the app.
3. Observe the queue-restore drain attempt log:
   `session-scoped RPC rejected ... not in memory (detached/reaped runtime; client should resume the stored session)`
4. After ~4 attempts the "排队消息未发送 / 多次发送失败" error banner appears.

### Expected behavior

- Queue recovery should **resume the stored session first** (the rejection message itself says "client should resume the stored session") and then submit, or
- if the target runtime is gone, clear the queued entry silently instead of surfacing a "message not sent / multiple send failures" error — the user has not lost anything and cannot fix it from the banner beyond manually retrying.

### Environment

- Hermes Agent v0.20.6 (2026.8.27), macOS 26.4.1, desktop app
- Profile-based install, `HERMES_HOME` per profile
- Relevant sources: `apps/desktop/src/app/session/hooks/use-background-queue-drain.ts`, `apps/desktop/src/store/composer-queue.ts` (`STORAGE_KEY`, `MAX_AUTO_DRAIN_ATTEMPTS = 4`), `tui_gateway/server.py` (session-scoped RPC rejection path)

salch-cred:
## Problem (#98015)

After an app restart all session runtimes are reaped. The composer queue persists in \localStorage\, so queued entries survive. On first drain attempt after restart, \
untimeIdByStoredSessionIdRef\ has no entry yet (runtime is reconnecting), so \submitText\ returns \alse\. After 4 fast retries the hook shows the \queueStuckTitle\ / \queueStuckBody\ banner — even though the message is safely in the queue and will send once the runtime is ready.

## Fix

Clear \drainFailuresRef\ on mount. Each fresh process starts with a clean slate so the banner never fires because of reaped runtimes from a previous process. The banner still fires for genuine send failures in the same process lifetime.

Fixes #98015

### DIFF
--- a/apps/desktop/src/app/session/hooks/use-background-queue-drain.ts
+++ b/apps/desktop/src/app/session/hooks/use-background-queue-drain.ts
@@ -53,6 +53,15 @@ export function useBackgroundQueueDrain({
   const retryTimersRef = useRef<number[]>([])
   const [retryTick, setRetryTick] = useState(0)
 
+  // On mount, clear any failure counters carried over from a previous process.
+  // After an app restart all session runtimes are reaped; the backend is
+  // starting up fresh, so previous failure counts that tripped MAX_AUTO_DRAIN
+  // do NOT represent real send failures — they were rejections against a dead
+  // runtime.  Resetting on mount gives each entry a clean slate so the banner
+  // is not shown on the first interaction after a restart (#98015).
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => { drainFailuresRef.current.clear() }, [])
+
   // eslint-disable-next-line no-restricted-syntax -- legitimate non-atom ref write (see eslint rule comment)
   useEffect(() => {
     submitTextRef.current = submitText


Enough1122:

**Overall:** Minimal fix for misleading queue-drain banner after app restart (#98015). Clearing `drainFailuresRef` on mount gives each session a clean slate after runtimes are reaped.

**Correctness:**
- `apps/desktop/src/app/session/hooks/use-background-queue-drain.ts:15` — `useEffect(() => { drainFailuresRef.current.clear() }, [])` correctly runs once on mount. `eslint-disable` for exhaustive-deps is appropriate (mount-only semantics).
- No dependency on `drainFailuresRef` identity (ref stable), so empty deps is safe.

**Non-blocking nit:**
- If future logic adds persistence for drain failures, this clear-on-mount would mask it — worth adding a comment that this is intentional for in-memory map only.

**Verdict:** LGTM.

PR #98579, opened 2026-08-30
fix(desktop): clamp oversized pastes and cap persisted composer drafts

### LINKED ISSUE #98562: [Bug]: macOS Desktop — pasting very long text freezes the app; reopen re-freezes until the chat is deleted
## Bug Description

Pasting a very long text into the Hermes Desktop composer **freezes the whole app**. The user has to Force Quit. On reopen, Hermes **immediately freezes again** because it restores that chat/draft. The only recovery is a race: delete the offending chat from the sidebar in under about a second, before hydrate locks the renderer.

This is a launch-blocking / data-loss-adjacent bug: one oversized paste makes the app unusable until that session is deleted.

## Steps to Reproduce

1. Open Hermes Desktop on macOS. Focus the main composer.
2. Paste a very large text blob (hundreds of KB to a few MB: a long log, dumped article, huge stack trace, etc.).
3. Observe: UI freezes (no scroll, no click, no quit from inside the app).
4. Force Quit Hermes.
5. Reopen Hermes Desktop.
6. Actual: it freezes again as soon as the same chat/draft hydrates.
7. Workaround: on reopen, immediately delete that chat in the sidebar (sub-second) before resume finishes. Any slower and it locks again.

## Expected Behavior

- Paste of large text never freezes the app. Excess is truncated, attached as a file, or inserted as plain text off the main thread / in chunks.
- Reopening Hermes always yields a usable window within a second, even if a session contains a huge draft or huge user message. Oversized content can stub (“message too large — load”) instead of locking the renderer.
- Recovery must not require a speed-run delete.

## Actual Behavior

- Large paste → full app freeze.
- Quit → reopen → freeze again on the same chat.
- Only fix: delete that chat in under ~1s on the next launch.

## Affected Component

Hermes Desktop (`apps/desktop`) composer + session resume on macOS. Template dropdown has no Desktop option.

- `apps/desktop/src/app/chat/composer/index.tsx` — `handlePaste`
- `apps/desktop/src/app/chat/composer/url-refs.ts` — `linkifyUrls`
- `apps/desktop/src/app/chat/composer/rich-editor.ts` — `insertComposerContentsAtCaret`, `appendComposerContents`, `renderComposerContents`, `appendTextWithBreaks`
- `apps/desktop/src/store/composer.ts` — `persistDraftTexts` / `stashSessionDraft` (`hermes:composer-drafts:v3`)
- Session transcript hydrate of a huge sent user message (same freeze-on-reopen if the paste was submitted)

## Debug Report

N/A — renderer lock, no useful traceback before Force Quit. Do not require `hermes debug share`.

## Operating System

macOS

## Python Version

N/A (Desktop UI)

## Hermes Version

Hermes Desktop, macOS. Still present on current `main` (grounded against `66666f6e2eca`).

UI-only freeze. No debug paste.

## Root Cause Analysis

Paste is fully synchronous on the UI thread, with **no size cap**, then the result is **persisted and re-hydrated on the next launch**.

| What | File | Symbol / detail |
|------|------|-----------------|
| Paste handler | `apps/desktop/src/app/chat/composer/index.tsx` ~L496–561 | `preventDefault`, then `insertComposerContentsAtCaret(…, pathifyRefs(linkifyUrls(pastedText)), scope)` on the **entire** clipboard string |
| URL rewrite over the whole blob | `apps/desktop/src/app/chat/composer/url-refs.ts` `linkifyUrls` | `text.matchAll(REF_RE)` + `text.matchAll(URL_RE)` — unbounded |
| Chip/DOM build | `apps/desktop/src/app/chat/composer/rich-editor.ts` `appendComposerContents` / `appendTextWithBreaks` | One DOM node per chip + per line. A 20k-line paste = tens of thousands of nodes on the main thread |
| Known O(n)/O(n²) | `index.tsx` ~L421; `rich-editor.ts` ~L291 | Comments already admit `composerPlainText` is O(n) and Chromium insert is ~O(n²) on large multiline blobs — paste still does the expensive path with no cap |
| Draft persist, no byte cap | `apps/desktop/src/store/composer.ts` | `persistDraftTexts()` writes full draft text to `localStorage` (`SESSION_DRAFTS_STORAGE_KEY`, `MAX_PERSISTED_DRAFTS = 50` count-only) |
| Reopen freeze | same + session resume | Last session / draft hydrates via `renderComposerContents` (full chip DOM) and/or the transcript renderer. That is why Force Quit is not enough and why deleting the **chat** in &lt;1s is the only workaround |

```ts
// handlePaste — no max length
const pastedText = sanitizeComposerInput(event.clipboardData.getData('text').trim())
event.preventDefault()
insertComposerContentsAtCaret(event.currentTarget, pathifyRefs(linkifyUrls(pastedText)), scope)
```

If the user also **sends** the blob, the session store keeps a huge user message; resume re-renders it and the freeze follows the chat rather than only the composer draft. Either persist path matches “delete the offending chat.”

## Proposed Fix

1. **Hard cap on paste in the composer** (e.g. 100k characters or 256 KiB). Over the cap: do not run `linkifyUrls` / `pathifyRefs` / chip DOM. Insert a truncated plain-text prefix **or** attach the rest as a file, and tell the user. Never `preventDefault` into a synchronous full-document rewrite of megabytes.
2. **Chunk / idle-callback** any DOM insert that remains, so one paste cannot block input for seconds.
3. **Cap persisted drafts.** Do not `localStorage.setItem` a draft over the cap; drop or stub it. Persist must never be able to poison the next cold launch.
4. **Safe resume.** On session activate, if draft or any message exceeds the cap, render a stub (“Large paste omitted — load / delete”) instead of building the full contenteditable/markdown tree before first paint. The window must stay interactive so the user can delete the chat without a 1-second race.
5. Transcript side: huge user/assistant messages need clamp + virtualize (existing sticky/list budget), not a single markdown pass over megabytes.

```ts
const PASTE_CHAR_CAP = 100_000
if (pastedText.length > PASTE_CHAR_CAP) {
  event.preventDefault()
  insertPlainText(pastedText.slice(0, PASTE_CHAR_CAP))
  // optional: attach overflow as a file; never linkify/chip the tail
  return
}
```

## Tests

- [ ] `handlePaste`: 2 MB / 100k-line clipboard does not call `linkifyUrls` on the full string; UI remains responsive (fake timers / does not occupy the thread &gt; 50ms).
- [ ] `persistDraftTexts`: drafts over the cap are not written; next load does not rehydrate megabytes.
- [ ] Session resume fixture with a 2 MB user message: shell (sidebar, composer, quit) paints and is clickable; the huge body is stubbed.
- [ ] Delete-session still works when the oversized session is the last-active one (no race required).
- [ ] Small pastes (normal messages, URLs, `@url:` chips) unchanged.

## Acceptance criteria

- [ ] Pasting very long text never freezes Hermes Desktop. Worst case: truncate/attach + a visible notice.
- [ ] Force Quit is not required.
- [ ] Reopen after an oversized paste/send is usable without deleting the chat in under a second.
- [ ] Deleting the chat is a normal action, not the recovery mechanism.
- [ ] Normal-size paste behavior (chips, URLs, undo) unchanged.

## Priority

P0 — one paste makes the app unrecoverable across restarts except by a sub-second sidebar delete.

liuhao1024:
## What does this PR do?

Pasting a multi-megabyte blob into the desktop composer froze the whole app (#98562): the paste handler ran `linkifyUrls`/`pathifyRefs` and the chip-DOM build synchronously over the entire blob on the UI thread, and the resulting draft was persisted to `localStorage` — so the next launch rehydrated it into another full DOM rebuild and re-froze the renderer before the window could paint. The only recovery was a sub-second race to delete the chat before hydrate locked the renderer again.

This PR bounds both ends of that pipeline with a single shared ceiling, `MAX_COMPOSER_TEXT_CHARS = 100_000`:

- **Paste clamp**: both paste entry points — the focused editor's `handlePaste` and the window-level `routeClipboardToComposer` (paste on non-editable chrome) — now share a new `preparePastedText()`. A paste over the cap inserts a plain-text prefix: no linkify, no pathify, no chip DOM. Normal-sized pastes keep the existing chipping behavior byte-for-byte.
- **Persist cap**: `persistDraftTexts()` drops drafts over the same cap instead of writing them to `localStorage`, so an oversized draft can never poison the next cold launch. The in-memory draft survives for the current session, so nothing is lost until the app restarts.

The issue also proposes chunked DOM insertion, safe-resume stubs, and transcript virtualization — those are larger follow-ups; this PR lands the minimal ceiling that removes the hard freeze and the launch-locking persistence path.

## Related Issue

Fixes #98562

## Type of Change

## Changes Made

- `apps/desktop/src/store/composer.ts` — add exported `MAX_COMPOSER_TEXT_CHARS = 100_000`; `persistDraftTexts()` now filters out drafts over the cap before writing `localStorage`.
- `apps/desktop/src/app/chat/composer/paste-to-focus.ts` — add `preparePastedText()` (clamp + shared chip prep); `routeClipboardToComposer()` routes through it.
- `apps/desktop/src/app/chat/composer/index.tsx` — `handlePaste` inserts `preparePastedText(pastedText)` instead of inlining `pathifyRefs(linkifyUrls(...))` (imports for the now-unused symbols removed).
- `apps/desktop/src/app/chat/composer/paste-cap.test.ts` — new: clamp at cap+1, chipping at exactly cap, plain-prefix (no `@url:`) for oversized input, and window-paste routing through the clamp.
- `apps/desktop/src/store/composer.test.ts` — new cases: oversized draft not persisted, kept in memory, and a draft exactly at the cap still persisted.

## How to Test

1. `cd apps/desktop && npx vitest run src/app/chat/composer/ src/store/composer.test.ts` — should pass: 49 files, 412 tests (Observed result: all green, including the 5 new paste-cap tests and 4 new draft-cap tests).
2. `npm run typecheck` and `npm run lint` in `apps/desktop` — should pass with no errors.
3. Manual: focus the composer, paste a >100k-character blob — should insert the first 100k characters as plain text within the same frame (no freeze, links in the prefix stay bare text). Paste a normal-sized text containing a URL — should still chip as `@url:`. Reload the app with an oversized draft stashed — should boot to an interactive window with an empty draft instead of freezing.

### Code

- [x] I've run `pytest tests/ -q` and all tests pass (Python suite untouched — this PR is renderer/TS-only; validated via the desktop vitest suite above)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (behavior comment documents the cap where it is defined)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) (paste path is platform-neutral renderer logic; cap applies identically everywhere)

### DIFF
--- a/apps/desktop/src/app/chat/composer/index.tsx
+++ b/apps/desktop/src/app/chat/composer/index.tsx
@@ -63,7 +63,8 @@ import { useComposerMicroActions } from './hooks/use-micro-actions'
 import { useSlashCompletions } from './hooks/use-slash-completions'
 import { useSessionStatusPresence } from './hooks/use-status-presence'
 import { ActionBadges } from './micro-actions'
-import { chipTypedPathOnSpace, pathifyRefs } from './path-refs'
+import { preparePastedText } from './paste-to-focus'
+import { chipTypedPathOnSpace } from './path-refs'
 import { QueuePanel } from './queue-panel'
 import {
   beginComposerComposition,
@@ -83,7 +84,7 @@ import { ComposerTriggerPopover } from './trigger-popover'
 import type { ChatBarProps } from './types'
 import { isRedoShortcut, isUndoShortcut } from './undo-history'
 import { UrlDialog } from './url-dialog'
-import { chipTypedUrlOnSpace, linkifyUrls } from './url-refs'
+import { chipTypedUrlOnSpace } from './url-refs'
 import { VoiceActivity, VoicePlaybackActivity } from './voice-activity'
 
 export function ChatBar({
@@ -558,7 +559,7 @@ export function ChatBar({
     const scope = openDirectiveScope(event.currentTarget)
 
     recordUndoPoint()
-    insertComposerContentsAtCaret(event.currentTarget, pathifyRefs(linkifyUrls(pastedText)), scope)
+    insertComposerContentsAtCaret(event.currentTarget, preparePastedText(pastedText), scope)
     scheduleFlushEditorToDraft(event.currentTarget)
   }
 
--- a/apps/desktop/src/app/chat/composer/paste-cap.test.ts
+++ b/apps/desktop/src/app/chat/composer/paste-cap.test.ts
@@ -1,0 +1,79 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+import { MAX_COMPOSER_TEXT_CHARS } from '@/store/composer'
+
+import { onComposerInsertRequest } from './focus'
+import { preparePastedText, routeClipboardToComposer } from './paste-to-focus'
+
+/** Minimal DataTransfer stand-in: text/plain only (the cap path never sees files). */
+function clipboard({ text = '' } = {}): DataTransfer {
+  return {
+    files: { item: () => null, length: 0 },
+    getData: (type: string) => (type === 'text' || type === 'text/plain' ? text : ''),
+    items: []
+  } as unknown as DataTransfer
+}
+
+/** The bus defers dispatch a macrotask; flush it. */
+const flushBus = () => new Promise(resolve => setTimeout(resolve, 1))
+
+afterEach(() => {
+  document.body.replaceChildren()
+  vi.restoreAllMocks()
+})
+
+// #98562: a multi-megabyte paste used to run linkifyUrls/pathifyRefs and the
+// chip-DOM build synchronously over the whole blob on the UI thread, freezing
+// the app — and the resulting draft rehydrated on the next launch to freeze
+// it again. preparePastedText clamps oversized input to a plain-text prefix
+// instead; normal-sized pastes keep the chipping behavior unchanged.
+describe('preparePastedText', () => {
+  it('chips links in a normal-sized paste', () => {
+    expect(preparePastedText('see https://example.com/docs')).toContain('@url:')
+  })
+
+  it('clamps an oversized paste to a plain-text prefix without chipping', () => {
+    const oversized = `${'a'.repeat(MAX_COMPOSER_TEXT_CHARS)}https://example.com/never-reached`
+
+    const prepared = preparePastedText(oversized)
+
+    expect(prepared).toBe('a'.repeat(MAX_COMPOSER_TEXT_CHARS))
+    expect(prepared).not.toContain('@url:')
+  })
+
+  it('still chips a paste exactly at the cap', () => {
+    const tail = ' https://example.com/x'
+    const atCap = `${'a'.repeat(MAX_COMPOSER_TEXT_CHARS - tail.length)}${tail}`
+
+    expect(atCap).toHaveLength(MAX_COMPOSER_TEXT_CHARS)
+    expect(preparePastedText(atCap)).toContain('@url:')
+  })
+})
+
+describe('routeClipboardToComposer oversized paste', () => {
+  it('inserts the clamped plain-text prefix instead of the full blob', async () => {
+    const inserts: string[] = []
+    const off = onComposerInsertRequest(({ text }) => inserts.push(text))
+
+    const oversized = `${'b'.repeat(MAX_COMPOSER_TEXT_CHARS)}https://example.com/later`
+
+    expect(routeClipboardToComposer(clipboard({ text: oversized }))).toBe(true)
+    await flushBus()
+    off()
+
+    expect(inserts).toHaveLength(1)
+    expect(inserts[0]).toBe('b'.repeat(MAX_COMPOSER_TEXT_CHARS))
+    expect(inserts[0]).not.toContain('@url:')
+  })
+
+  it('keeps link chipping for a normal-sized window paste', async () => {
+    const inserts: string[] = []
+    const off = onComposerInsertRequest(({ text }) => inserts.push(text))
+
+    routeClipboardToComposer(clipboard({ text: 'see https://example.com/docs' }))
+    await flushBus()
+    off()
+
+    expect(inserts[0]).toContain('@url:')
+  })
+})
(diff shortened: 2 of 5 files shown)


Enough1122:

**Overall:** Solid defense against multi-MB paste freeze (#98562). Clamping to `100_000` chars before DOM chipping and dropping oversized drafts from `localStorage` prevents re-freeze on next launch.

**Correctness:**
- `apps/desktop/src/app/chat/composer/paste-to-focus.ts:139-145` — `preparePastedText` centralizes both paste entry points (focused editor + window-level route) so ceiling cannot be bypassed. Plain-text prefix slice avoids synchronous `linkifyUrls`/`pathifyRefs` over huge blob.
- `apps/desktop/src/store/composer.ts:231` — persistence filter `length <= MAX_COMPOSER_TEXT_CHARS` plus `caps` test `300 +` lines covers edge cases (at-cap kept, over-cap dropped, in-memory retained for current session).
- Tests in `paste-cap.test.ts` and `composer.test.ts` are precise (isolation via `preparePastedText` unit + bus flush).

**Non-blocking nits:**
- Truncation is silent — user may not realize paste was clipped. Consider a transient notice when `text.length > MAX` (UX, not blocking).
- `MAX_COMPOSER_TEXT_CHARS = 100_000` is generous for `localStorage` (5 MB typical limit); 50 drafts × 100k ≈ 5 MB — borderline. `MAX_PERSISTED_DRAFTS=50` slice mitigates, but worth monitoring.

**Verdict:** LGTM.

liuhao1024:
Thanks for the review — both nits considered:

- **Silent truncation notice:** agreed it's a real UX gap, but it's an enhancement beyond this PR's freeze fix (the cap exists to stop the freeze; teaching the user about clipping is a separate concern with its own copy/i18n decisions). I'd rather keep this diff scoped to the regression; happy to pick it up in a follow-up if there's interest.
- **`100_000` chars vs the ~5 MB `localStorage` ceiling:** fair math — 50 drafts at the theoretical max would land right at the edge. In practice drafts average orders of magnitude below the cap and `MAX_PERSISTED_DRAFTS` slicing bounds the worst case, so I've left the constants as-is for now; noted as something to revisit if quota-exceeded reports ever surface.

PR #98582, opened 2026-08-30
feat(kanban): add non_dispatchable_assignees gate for human-owner lanes

### LINKED ISSUE #98560: Kanban: no per-assignee dispatch exemption for 'ready' — only kanban.review_dispatch gates 'review', so a no-model profile still gets spawned and crashes
### Summary

### Concrete failure

### Request

### Environment

kokhlo:
## Summary

Adds `kanban.non_dispatchable_assignees: [<profile>, ...]` to opt a real Hermes profile out of autonomous dispatch entirely — the per-assignee counterpart of `kanban.review_dispatch`. 

A **human-owner lane** is a real profile (created for identity and shared memory) with no model configured: `profile_exists` returns `True`, so the existing nonexistent-profile guard cannot catch it. Without this list, the dispatcher spawns a worker that exits immediately on the missing model, and the block-recurrence breaker files a card that was waiting on a person as `blocked`.

The request in #98560 described 22 such stuck cards before the reporter deployed an external job racing the dispatch tick — a strong signal this is a real fleet pain.

## Changes

- Add `kanban.non_dispatchable_assignees` config; normalize entries with `normalize_profile_name` so title-cased dashboard labels match the lowercase lane.
- Resolve the list once per tick alongside `profile_exists`, consult it in both ready and review loops, and bucket matches into `skipped_nonspawnable` (steady-state, not operator-actionable).
- Apply the exemption to `has_spawnable_ready` / `has_spawnable_review` so health telemetry sees a queue full of exempt lanes as "correctly idle", not "dispatcher stuck".
- Tests: exempt lane is skipped in both columns; other assignees spawn; title-cased config matches lowercase lane; telemetry behaves correctly.

## Testing

```sh
uv run pytest -xv tests/hermes_cli/test_kanban_review_lifecycle.py tests/hermes_cli/test_kanban_default_assignee.py
# 23 passed in 2.38s
```

Full kanban suite passing on my fork — this change touches the ready/review dispatch paths, the two busiest hot loops, so CI's broader platform matrix will verify there's no cross-platform drift.

Closes #98560

### DIFF
--- a/hermes_cli/config_defaults.py
+++ b/hermes_cli/config_defaults.py
@@ -2811,6 +2811,12 @@ DEFAULT_CONFIG = {
         # the assigned profile with the bundled sdlc-review skill. Disable for
         # boards where every review is performed manually from the dashboard.
         "review_dispatch": True,
+        # Assignees the dispatcher must never spawn workers for, even though
+        # they are real Hermes profiles — e.g. a human-owner lane created for
+        # identity/shared memory with no model configured. Listed lanes are
+        # treated like control-plane lanes: their cards are skipped as
+        # non-spawnable in both the ready and review columns.
+        "non_dispatchable_assignees": [],
         # Seconds between dispatcher ticks (idle or not). Lower = snappier
         # pickup of newly-ready tasks; higher = less SQL pressure.
         "dispatch_interval_seconds": 60,
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -9569,12 +9569,18 @@ def has_spawnable_ready(conn: sqlite3.Connection) -> bool:
     ).fetchall()
     if not rows:
         return False
+    # ``kanban.non_dispatchable_assignees`` lanes are never spawned, so
+    # they are "correctly idle" — otherwise a human-owner lane parked in
+    # ready would fire a permanent false "dispatcher stuck" warning.
+    _exempt = non_dispatchable_assignees()
     try:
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** Well-scoped feature: `kanban.non_dispatchable_assignees` gate for human-owner lanes that are real profiles but must never be auto-dispatched. Applies to both ready and review columns.

**Correctness:**
- `hermes_cli/kanban_db.py:60-101` — `non_dispatchable_assignees()` normalizes via `normalize_profile_name` with fallback to `lower()`, handles `str`/`list`/tuple, strips, ignores empties. Case-insensitive matching matches dashboard title-cased labels.
- `hermes_cli/kanban_db.py:30,46,112-128` — exemption in `has_spawnable_ready/review()` prevents false "dispatcher stuck" telemetry; exemption in `_any_spawnable_review` dispatch loop before per-profile caps is correctly ordered.
- Tests `test_non_dispatchable_assignee_*` cover both columns, telemetry, toggle-off proof, and case-insensitivity.

**Non-blocking nits:**
- `load_config()` called per tick / per `has_spawnable_*` — cache via `_LOAD_CONFIG_CACHE` mitigates, but consider passing down `_non_dispatchable` set to avoid re-resolving in `has_spawnable_*` if called in hot path (minor).
- Docs in `config_defaults.py` clearly state "treated like control-plane lanes" — good.

**Verdict:** LGTM.

PR #98585, opened 2026-08-30
fix(state): reduce _READ_POOL_MAX 8→4 to halve peak state.db file-descriptor budget

### LINKED ISSUE #98573: long-lived gateway holds ~20 state.db handles per process; per-thread read connections are never released
## Problem

A long-running gateway process holds ~20 open file handles to `state.db` (plus its
`-wal` / `-shm` sidecars) for the entire life of the process. The handle count grows
linearly with the worker-thread count and never shrinks. On macOS (default
`RLIMIT_NOFILE` soft = 256), combined with httpx sockets and terminal subprocess
pipes, this exhausts the fd budget and unrelated code paths start failing with
`OSError: [Errno 24] Too many open files`.

## Evidence (lsof on a live gateway)

```
$ lsof -p 812 | grep -iE '\.db'
20  /Users/lining/.hermes/state.db
12  /Users/lining/.hermes/state.db-wal
 2  /Users/lining/.hermes/state.db-shm
```

Same process had 19 threads at that moment (`ps -M 812` → 20 lines incl. header),
i.e. 19 read connections + 1 writer connection:

```
$ lsof -p 812 | grep -icE '\.db'    # 34
$ ps -M 812 | wc -l                 # 20 (19 threads + header)
```

## Root cause

`hermes_state.py::SessionDB` opens one **read-only connection per thread** and pins it:

- `self._read_local = threading.local()`  — per-thread read connection
- `_get_read_conn()` stores the connection in `_read_local.conn` and also in
  `self._read_conns` (a strong-reference set so short-lived reader threads don't
  leak tracked fds)
- `close()` drains `_read_conns`, but a **long-lived gateway never calls `close()`**
  during normal operation

The gateway runs agents on a `ThreadPoolExecutor` whose threads are pooled and never
exit, so every worker thread that ever touches a recall/browse read path opens its
own `state.db` read connection and holds it forever. Thread pool grows → handle count
grows → never reclaimed until process exit. The writer connection (`self._conn`) is a
singleton, but the read connections are effectively unbounded (one per distinct
worker thread).

## Impact

- On macOS the default soft `RLIMIT_NOFILE` is 256 (inherited by launchd-spawned
  LaunchAgents). ~20 SQLite handles + httpx/OpenAI sockets + terminal subprocess
  pipes pass 256 under load, producing `[Errno 24] Too many open files` in unrelated
  code paths (observed in `tools/terminal_tool.py` cleanup thread).
- The same per-thread connection design means fd footprint scales with thread-pool
  size — a latent leak on any platform, not just macOS.

## Environment

- macOS 26.3.2 (arm64), Python 3.11
- Hermes gateway running as a launchd LaunchAgent (long-lived)

## Suggested direction

- Bound/reuse read connections instead of one-per-thread-unbounded (e.g. a small
  connection pool, or fall back to the writer connection for reads once thread count
  exceeds a cap), or
- Release idle per-thread read connections after an inactivity timeout.

salch-cred:
## Problem (#98573)

A long-lived gateway holds ≥2 \SessionDB\ instances. With \_READ_POOL_MAX=8\, peak \state.db\ fd usage is \2 × 8 × 2 + 4 = 36\ fds (each connection holds \state.db\ + \state.db-wal\). Combined with httpx sockets and subprocess pipes this exhausts the macOS 256-fd soft \RLIMIT_NOFILE\, causing \OSError: Too many open files\ in unrelated code paths on a healthy process the supervisor never restarts.

## Fix

Reduce \_READ_POOL_MAX\ from \8\ to \4\. Peak drops to 20 fds — a 45% reduction. WAL read-concurrency is shared via the page cache; the 4th+ simultaneous reader is uncommon in practice. The fallback overflow path (degrade to the locked writer connection) is already implemented and measured by \_read_permit_exhausted\.

Fixes #98573

### DIFF
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -382,7 +382,23 @@ _READ_OPEN_RETRY_SECONDS = 60.0
 # close. Once permits are gone the read path degrades to the locked writer
 # connection instead of opening more descriptors — slower under load, which is
 # the correct trade against a process-wide wedge the supervisor cannot see.
-_READ_POOL_MAX = 8
+# Read-connection pool ceiling per SessionDB instance.
+#
+# Each open connection holds two file descriptors (state.db + state.db-wal).
+# A gateway process runs at least two SessionDB instances (tui_gateway +
+# gateway/run.py shared db), so the peak fd budget is:
+#
+#   instances × pool_max × 2 fds/conn + 2 writer fds/instance
+#
+# With pool_max=8: 2 × 8 × 2 + 2 × 2 = 36 fds from state.db alone.
+# On macOS the default soft RLIMIT_NOFILE is 256; combined with httpx
+# sockets and subprocess pipes a long-lived gateway with many worker threads
+# exhausted the limit (#98573).
+#
+# Reduce to 4: 2 × 4 × 2 + 2 × 2 = 20 fds — a 45 % reduction in the
+# worst case, with minimal read latency regression (WAL readers share pages
+# in the page cache; the 4th+ concurrent reader is uncommon in practice).
+_READ_POOL_MAX = 4
 
 # Import-time snapshot used by _default_db_path() to detect a deliberately
 # re-pointed DEFAULT_DB_PATH (tests monkeypatch the constant directly).


Enough1122:

**Overall:** One-line tuning: `_READ_POOL_MAX 8 → 4` to halve peak `state.db` fd budget (36 → 20) for macOS `RLIMIT_NOFILE 256` exhaustion (#98573).

**Correctness:**
- `hermes_state.py:10-26` — comment quantifies math: `instances × pool_max × 2 + 2 writer fds`. Reduction 45 % with justification (WAL readers share page cache; 4th+ concurrent reader uncommon).
- No behavioral change beyond pool size; degrades to locked writer when permits exhausted — documented graceful fallback.

**Non-blocking nit:**
- Consider exposing as configurable (e.g. `state.read_pool_max`) for operators with raised ulimit who prefer latency over fds. Not required for this fix.

**Verdict:** LGTM — concise, well-rationalized.

PR #98586, opened 2026-08-30
fix: surface provider queueing/rate-limit backoff in live UI status

weihan93751:
## Summary
When a provider returns 429 (queueing/rate-limited — e.g. coding-plan Token Plan), the retry loop silently backs off 5–120s while CLI/TUI/Desktop show an unexplained frozen spinner. The wait was only buffered (`_buffer_status`) and dropped on eventual success, so users could not tell a queueing provider from a hung turn.

This emits the existing `_emit_wait_notice` (→ `thinking.delta` → desktop provider-wait row) at three points:
- immediately when a 429/overload is classified (with attempt count + wait time)
- every ~30s during the invalid-response backoff heartbeat
- every ~30s during the exception-path backoff heartbeat

The notice text is generated by a new `_provider_wait_notice_text()` helper so the cross-surface contract (desktop `providerWaitText` prefix match) is unit-testable.

**Localization**: all user-facing wait/reconnect notices are now Chinese (the operator's language):
- `⏳ 等待 qwen3.8-flash — 对方排队/限流中，45 秒后重试（第 2/4 次）`
- `⚠ 无响应：对方 30 秒未返回 — 正在重新连接...`
- `⚠ 无输出：对方 30 秒未返回 — 正在重新连接...`
- `↻ 模型已返回推理但无最终答复 — 正在请它继续（1/3）`
- plus the existing 30s no-output / TTFB reconnect lines

**Desktop filter**: `providerWaitText` previously matched only English prefixes (`⏳ waiting on` / `⚠ no output` / `↻ model returned`), which would silently drop the localized text. Extended to also match `⏳ 等待` / `⚠ 无输出|无响应` / `↻ 模型已返回`.

## Test Plan
- Python: `tests/run_agent/test_wait_state_visibility.py` — desktop-contract regex mirror (bilingual), roundtrip through `thinking_callback`, reconnect notice. `tests/agent/test_codex_ttfb_watchdog.py` updated for localized phrasing. 13/13 pass.
- Desktop: `status.test.tsx` adds a Chinese queueing-notice render case.

### DIFF
--- a/agent/chat_completion_helpers.py
+++ b/agent/chat_completion_helpers.py
@@ -1632,9 +1632,8 @@ def interruptible_api_call(agent, api_kwargs: dict):
                     elapsed=_elapsed,
                 )
                 agent._emit_wait_notice(
-                    f"⏳ waiting on {api_kwargs.get('model', 'the provider')} — "
-                    f"{int(_elapsed)}s with no response yet (provider may be slow "
-                    f"or overloaded{_recovery})"
+                    f"⏳ 等待 {api_kwargs.get('model', 'the provider')} — "
+                    f"已 {int(_elapsed)} 秒无响应（对方可能较慢或过载{_recovery}）"
                 )
             except Exception:
                 logger.debug("wait-notice construction failed", exc_info=True)
@@ -1682,8 +1681,7 @@ def interruptible_api_call(agent, api_kwargs: dict):
             except Exception:
                 pass
             agent._emit_wait_notice(
-                f"⚠ no response from provider in {int(_elapsed)}s — "
-                f"reconnecting..."
+                f"⚠ 无响应：对方 {int(_elapsed)} 秒未返回 — 正在重新连接..."
             )
             agent._touch_activity(
                 f"codex stream killed after {int(_elapsed)}s with no first byte"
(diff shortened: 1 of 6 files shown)


Enough1122:

**Overall:** UX improvement surfacing provider queueing/429 backoff in live status (CLI/TUI/Desktop `thinking.delta`). Previously silent backoff read as frozen turn.

**Correctness:**
- `agent/conversation_loop.py:58-76` — `_provider_wait_notice_text` keeps `⏳ 等待` prefix required by `providerWaitText` contract. Tests `test_backoff_wait_notice_matches_desktop_provider_wait_contract` guard this cross-surface invariant.
- Emits wait notice both immediately on 429 (`conversation_loop.py:115-123`) and every 30 s during sleep loop (`...:98-102, 144-147`), with activity fallback preserved.
- `apps/desktop/src/store/provider-wait.ts:192` regex extended to `等待|无(?:输出|响应)|模型已返回` alongside English — both locales pass filter.
- i18n strings in `chat_completion_helpers.py` switched to Chinese consistently.

**Non-blocking nits:**
- Localized strings may need downstream translation pipeline update if English UI is expected elsewhere — verify product i18n policy. Current change is intentional per PR title.
- Backoff loop still sleeps `0.2 s` increments — wait notice throttling `150 × 0.2 s = 30 s` is correct, but first notice after entering backoff waits 30 s; immediate notice covers the gap (good).

**Verdict:** LGTM.

PR #97014, opened 2026-08-28
fix(env): scope terminal config re-bridge to the true process profile

Bergmann89:
## What does this PR do?

In a multiplexed dashboard/gateway (one process serving every profile on the host), a **secondary** profile's `terminal.backend` could leak into the shared `os.environ` and silently override the **launch** profile's terminal backend. A profile configured for `terminal.backend: ssh` would end up running every command on the local host because a sibling profile with `backend: local` had polluted the shared environment.

The tell is an asymmetric env state on the launch session:

```
TERMINAL_ENV=local             # leaked from a sibling profile
TERMINAL_SSH_HOST=10.10.0.103  # still the launch profile's — untouched
```

A restart does not help: as soon as the sibling profile is touched again (a turn, a cron tick), the shared env is re-poisoned.

### Root cause

`env_loader._reapply_terminal_config_bridge()` guards *"only re-bridge config into the shared `os.environ` when this load is for the process's own profile"*:

```python
if Path(home_path).resolve() != _process_hermes_home().resolve():
    return
```

`_process_hermes_home()` delegated to `get_hermes_home()`, which **follows the context-local `set_hermes_home_override`** a per-request task installs. When a per-turn handler is scoped to a secondary profile (via `_config_profile_scope` / `set_hermes_home_override`) and triggers a `load_hermes_dotenv()` reload, **both sides of the comparison resolve to that secondary profile** — the guard passes, and the bridge runs `apply_terminal_config_to_env(env=None)` against the shared `os.environ`, writing the *wrong* profile's `terminal.backend`.

Because the secondary profile's `config.yaml` carries no `TERMINAL_SSH_*` keys, only `TERMINAL_ENV` is overwritten (explicit-keys-only bridge semantics), producing the asymmetric state above.

The helper was originally introduced for the config-cache key (where following the active home is correct) and later reused for the terminal bridge guard, where the override-following semantics are wrong.

### The fix

Delegate `_process_hermes_home()` to `get_process_hermes_home()`, the override-immune resolver built for exactly this — it reflects the scope the process was **launched** under, ignoring per-task overrides. Both call sites (the bridge guard and the secrets-cache reuse check in `_load_secrets_config`) want the true process home.

This is the correct boundary: in a multiplex process `os.environ` belongs to the launch profile. A secondary profile's terminal config must reach its own sessions through session-scoped mechanisms (see Related), **never** by mutating the shared process environment.

## Related Issue

[...]

PR #94206, opened 2026-08-24
fix(tools): scope the terminal env bridge one-shot to the Hermes home

liuhao1024:
## What does this PR do?

Under profile multiplexing (`gateway.multiplex_profiles: true`), `_ensure_terminal_env_bridged()`'s process-global one-shot flag let the first profile to run a terminal call permanently pin its `TERMINAL_*` selection into `os.environ`. With a docker-backend profile (e.g. `coder`) bridging before a local-backend one (e.g. `board-leonard`), every later profile inherited Docker — intermittent "sandbox flapping" depending on cron tick order at gateway boot.

This scopes the one-shot to the Hermes home (the reporter's suggested direction, verified in their fork): the last-bridged home is tracked alongside the flag, and a scope change re-bridges. On top of the reporter's patch it also unsets the `TERMINAL_*` keys the previous bridge introduced before re-bridging — without that, a profile whose config has no `terminal` section keeps the previous profile's backend forever through the `elif "TERMINAL_ENV" not in os.environ` branch, which is the leak's worst shape. Shell/.env-exported `TERMINAL_*` values are never bridge-owned and survive handoffs; within a single profile the one-shot semantics (and its optimization purpose) are unchanged.

## Related Issue

Fixes #94200
Fixes #98581

## Type of Change

- [x] 🐛 Bug fix

## Changes Made

- `tools/terminal_tool.py` (`_ensure_terminal_env_bridged`):
  - New module globals `_terminal_config_bridge_scope` (last-bridged `str(get_hermes_home())`) and `_terminal_config_bridge_keys` (the `TERMINAL_*` keys the last bridge introduced).
  - The one-shot guard now compares scope; on a scope change the bridge-introduced keys are popped from `os.environ` and the bridge re-runs for the incoming profile's config. Docstring documents the multiplexing failure mode.
- `tests/tools/test_terminal_env_bridge.py`:
  - The autouse reset fixture also resets the two new globals.
  - Four new tests: docker→local handoff re-bridges; a section-less profile no longer inherits the previous profile's docker selection (the `elif` branch path); one-shot within a single scope is preserved (single `apply_terminal_config_to_env` call across three `_get_env_config()` runs); shell-exported `TERMINAL_*` keys survive handoffs.

## How to Test

- Run: `.venv/bin/python -m pytest tests/tools/test_terminal_env_bridge.py tests/tools/test_docker_session_isolation.py tests/tools/test_terminal_degraded_mode.py -q`
[...]

Issue #94200, opened 2026-08-24
_ensure_terminal_env_bridged one-shot leaks profile-specific TERMINAL_ENV across multiplexed profiles


betaeter:
tools/terminal_tool.py

Under Hermes multiplexing (gateway.multiplex_profiles: true), _ensure_terminal_env_bridged() in tools/terminal_tool.py uses a process-global one-shot flag (_terminal_config_bridge_attempted) that prevents re-entry. The first profile to trigger a terminal call permanently sets TERMINAL_ENV in os.environ. If a profile with terminal.backend: docker (e.g. coder) runs before one with terminal.backend: local (e.g. board-leonard), all profiles inherit Docker — causing intermittent "sandbox flapping" depending on cron tick order at gateway boot.

Fix applied in our fork: Track last-bridged HERMES_HOME scope alongside the one-shot flag. Re-bridge when scope changes:

global _terminal_config_bridge_attempted, _terminal_config_bridge_scope
from hermes_constants import get_hermes_home
_current_scope = str(get_hermes_home())
if _terminal_config_bridge_attempted and _terminal_config_bridge_scope == _current_scope:
    return
_terminal_config_bridge_attempted = True
_terminal_config_bridge_scope = _current_scope

alt-glitch:
> *This was generated by AI during triage.*

Duplicate of #68559: both reports identify the process-global terminal environment bridge leaking the first multiplexed profile's backend into subsequent routed profiles.

PR #85084, opened 2026-08-13
fix: scope terminal backend per multiplexed profile

CannibalKush:
## Summary

Fix terminal backend and sandbox selection when one gateway serves multiple Hermes profiles.

## Problem

A multiplexed gateway runs several profile turns in the same process. Terminal configuration was previously resolved through process-wide `TERMINAL_*` state. When turns overlapped, one profile could inherit another profile's backend—for example, an `agent` profile configured for Docker could accidentally use the host-local backend.

This was both a correctness bug and a sandbox-boundary risk: the wrong backend can expose the host filesystem and the wrong persistent environment.

## What changed

- Added a context-local terminal configuration scope for multiplexed profile turns; overlapping turns no longer mutate or read each other's backend settings.
- Routed terminal, file, code-execution, and environment-hint resolution through the active profile's scoped configuration.
- Included profile identity in environment and persistent-sandbox keys so matching task/session IDs cannot cause cross-profile reuse.
- Sanitized profile-qualified task IDs before using them in Docker host paths, preventing invalid `-v` mount specifications while retaining collision-resistant identity.
- Preserved the existing process-environment behavior for single-profile deployments.
- Failed closed when a multiplexed profile's terminal configuration cannot be resolved, rather than silently falling back to another backend.

## Security and scope

This PR isolates **terminal backend selection and sandbox reuse** across multiplexed profiles. It does not claim to solve every profile-isolation issue.

Credential inheritance into child processes and Kanban workers remains tracked separately in [#82936](https://github.com/NousResearch/hermes-agent/issues/82936). That issue requires child-process environment scrubbing and profile-aware credential allowlisting; it is complementary to, and not closed by, this PR.

## Validation

- Targeted terminal/profile/Docker regression suite: **67 passed** in the provisioned PR environment.
- Prompt-builder regression suite: **61 passed, 1 skipped**.
- Docker session-isolation suite after the mount-path fix: **28 passed**.
- Combined prompt/profile checks after the mount-path fix: **64 passed, 1 skipped**.
- `py_compile` and `git diff --check` passed.
- Live Discord dogfood confirmed the routed `agent` profile executed inside Docker:
  - `pwd` → `/root`
  - `HOME=/root`
  - Linux container runtime
  - host Hermes config and directory not visible

No credentials or authentication material are included in this change.

PR #98589, opened 2026-08-30
fix(tools): re-bridge terminal env per profile home, not once per process

### LINKED ISSUE #98581: [Bug]: Unified dashboard serves non-default profiles with the wrong terminal backend (docker sandbox escape via terminal tool)
### Bug Description

**Hermes version:** v0.20.6 (2026.8.27), commit 2a598aad
**Also present on:** `main` (66666f6e)
**Platform:** Linux (Debian 13), rootless podman, remote gateway

When a single unified dashboard backend (`hermes dashboard` without `--isolated`) serves multiple profiles through its profile switcher, every non-default profile inherits the **launch profile's** `TERMINAL_ENV` when executing commands through the **terminal tool**. Profiles configured with `terminal.backend: docker` end up running commands **on the host** — a sandbox escape.

## Impact

Security: profiles intended to be sandboxed (docker, restricted volumes, no host access) run with **full host access** when served by the unified dashboard. In our setup this affects `web` and `homeassistant` (both `terminal.backend: docker`). The workaround is to run one isolated dashboard per profile (`--isolated`), which gives each its own process and its own `TERMINAL_ENV` — but that defeats the unified profile-switcher model.

### Steps to Reproduce

Setup:
- `hermes dashboard --host 127.0.0.1 --port 9119 --no-open --skip-build` (no `--isolated`) → unified mode, launch profile = `default`.
- `default` config: `terminal.backend: local`.
- `web` and `homeassistant` configs: `terminal.backend: docker` (with `docker_volumes`).
- Connect the desktop app (or dashboard) to the 9119 backend and switch to the `web` / `homeassistant` profile.

### Expected Behavior

- Each profile should execute commands in its own configured terminal backend (docker for `web`/`homeassistant`), regardless of which process serves it.

### Actual Behavior

- Ask the profile to run:
  `if [ -f /.dockerenv ] || grep -q docker /proc/1/cgroup 2>/dev/null; then echo "IN DOCKER"; else echo "ON HOST"; fi`
- Both `web` and `homeassistant` report **"ON HOST"** (running on the host), even though their configs select `terminal.backend: docker` and their persistent podman containers exist and are running.

### Affected Component

Tools (terminal, file ops, web, code execution, etc.)

_No response_

### Debug Report

```shell
Report       https://paste.rs/F6URF
 agent.log    https://dpaste.com/A5QJS29J9
 gateway.log  https://dpaste.com/8FTGVNJJ4
 gui.log      https://dpaste.com/4GWZZAQT6
```

### Operating System

Debian 13

### Python Version

_No response_

### Hermes Version

_No response_

```shell

```

`tools/terminal_tool.py` `_ensure_terminal_env_bridged()` (~line 1713) bridges `terminal.*` config into `TERMINAL_*` env vars **once per process**, guarded by a module-global flag:

```python
global _terminal_config_bridge_attempted
if _terminal_config_bridge_attempted:
    return
_terminal_config_bridge_attempted = True
...
apply_terminal_config_to_env(env=None, override=True)
```

`_get_env_config()` (~line 1760) then reads `TERMINAL_ENV` from `os.environ`. Because the flag is module-global (not keyed by profile), the bridge runs only in the first profile context the process sees — the launch profile (`default` → `TERMINAL_ENV=local`). When the unified dashboard later serves a non-default profile via `set_hermes_home_override` (ContextVar, `tui_gateway/server.py` `_profile_home` / `_profile_scoped`, ~lines 2357/2374), the terminal env is **not** re-resolved from that profile's config, so `TERMINAL_ENV` stays `local`.

This is the same bug class as #63141 / #54449 / #61115 / #65696 (processes falling back to the local backend despite `terminal.backend: docker`), reintroduced specifically for the **multi-profile unified dashboard** path: the one-shot bridge fixes single-profile processes but not per-profile scoping within one process.

## Status on `main` (66666f6e)

**Not fixed for the terminal tool.** The `_terminal_config_bridge_attempted` flag is still one-shot on `main`, and `terminal_tool.py` still reads `TERMINAL_ENV` from `os.environ` bridged once for the launch profile.

There is a **partial fix** in `hermes_cli/web_server.py` (~lines 16470-16510): a per-profile re-bridge via `_config_profile_scope(requested)` + `apply_terminal_config_to_env(env=env)`. However, it only covers the **embedded chat PTY** (`_make_tui_argv` / `build_subprocess_env`), **not the terminal tool** the agent uses to execute commands. The regression test `tests/hermes_cli/test_web_server_profile_unification.py` pins config/env/PTY scoping but does not exercise agent command execution.

### Proposed Fix (optional)

Make the terminal-env bridge **per-profile** instead of one-shot module-global: re-run `apply_terminal_config_to_env` (or re-resolve `TERMINAL_*`) whenever the active profile context changes (keyed by the `HERMES_HOME` ContextVar override), so each profile served by a unified backend gets its own `TERMINAL_ENV` / `TERMINAL_DOCKER_*` values from its own `config.yaml`. The existing per-profile re-bridge in `web_server.py` for the PTY path is the right pattern; it needs to be applied to the terminal tool's execution path as well.

liuhao1024:
## What does this PR do?

The terminal config→env bridge in `tools/terminal_tool.py` (`_ensure_terminal_env_bridged`) was guarded by a one-shot module-global flag, so the first profile context the process saw — the launch profile — latched its `TERMINAL_*` selection for the whole process. In a unified dashboard backend (`hermes dashboard` without `--isolated`), every non-default profile served through the profile switcher therefore executed terminal-tool commands with the **launch profile's** terminal backend. Profiles configured with `terminal.backend: docker` ran commands **on the host** — a sandbox escape (#98581).

This keys the bridge guard by the **active Hermes home** (context-local profile override → `HERMES_HOME` env → platform default, via the existing `hermes_home_key()` helper) instead of a bare bool:

- when the active home changes, the `TERMINAL_*` vars the previous home's bridge (or the launch-time CLI bridge) explicitly **owned** — computed with the existing `terminal_config_owned_env_vars()` helper — are dropped before re-bridging, so a stale `TERMINAL_ENV=local` from the launch profile can never leak into a profile whose config selects docker;
- the re-bridge then resolves the new home's `terminal.*` config (`read_raw_config`/`apply_terminal_config_to_env` already resolve through `get_hermes_home()`, which honors the context-local override);
- within one home the bridge still runs at most once, and a failed attempt is still recorded for that home (same waste-avoidance as the original one-shot flag);
- a lock keeps the pop→bridge→record sequence atomic across threads.

This mirrors the per-profile re-bridge `hermes_cli/web_server.py` already performs for the embedded chat PTY (`_config_profile_scope` + `apply_terminal_config_to_env`); this PR applies the same pattern to the terminal-tool execution path the agent itself uses. Every caller of `_ensure_terminal_env_bridged` (`_get_env_config`, `_session_isolation_enabled`, `_docker_persistent_profile_scoped`) benefits.

## Related Issue

Fixes #98581

## Type of Change

- [x] 🔒 Security fix

## Changes Made

- `tools/terminal_tool.py` — replaced the one-shot `_terminal_config_bridge_attempted` bool with `_terminal_config_bridge_state` (`{home, owned}`); `_ensure_terminal_env_bridged` now re-runs when the active Hermes home changes, dropping the previous home's explicitly-owned `TERMINAL_*` vars first (fail-closed for sandboxed profiles).
- `tests/tools/test_terminal_env_bridge.py` — updated the autouse fixture for the new state shape; added two regressions: profile switch serves each profile's own backend (local→docker→local round-trip) and a launch-owned `TERMINAL_ENV` selection does not survive a switch to a profile without a terminal section.
- `tests/tools/test_docker_session_isolation.py`, `tests/tools/test_terminal_degraded_mode.py` — updated fixtures that pinned the old flag to pin the new per-home state instead (no behavior change for those suites).

## How to Test

1. Create two profiles: a launch profile with `terminal.backend: local` and a second with `terminal.backend: docker` (+ `docker_image`).
2. Start a unified backend: `hermes dashboard --no-open --skip-build`.
3. Switch to the docker profile and ask the agent to run: `if [ -f /.dockerenv ]; then echo IN DOCKER; else echo ON HOST; fi`.
   - Before: reports `ON HOST` (the launch profile's backend). Observed result after this PR: `IN DOCKER`, and `TERMINAL_ENV=docker` is bridged from that profile's config.yaml; switching back to the launch profile restores `local`.
4. Unit tests: `pytest tests/tools/ -k terminal -q` → 294 passed, 3 skipped; `pytest tests/tools/test_terminal_env_bridge.py tests/tools/test_docker_session_isolation.py tests/tools/test_terminal_degraded_mode.py tests/hermes_cli/test_web_server_profile_unification.py -q` → all pass (includes the two new regressions).

### Code

- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, etc.)
- [x] I've tested on my platform: macOS 15 (darwin 25.4.0, arm64)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (docstrings updated in place)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — or N/A (env-var logic is platform-neutral; path keys normalized by `hermes_home_key` via `os.path.normcase`)

### DIFF
--- a/tests/tools/test_docker_session_isolation.py
+++ b/tests/tools/test_docker_session_isolation.py
@@ -44,8 +44,14 @@ def _clean_state(monkeypatch):
     with terminal_tool._session_cwd_lock:
         before_cwd = dict(terminal_tool._session_cwd)
         terminal_tool._session_cwd.clear()
-    # The config→env bridge is one-shot; mark it done so tests control env vars.
-    monkeypatch.setattr(terminal_tool, "_terminal_config_bridge_attempted", True)
+    # The config→env bridge runs once per Hermes home; mark it done for the
+    # active home so tests control env vars directly.
+    from hermes_constants import hermes_home_key
+    monkeypatch.setattr(
+        terminal_tool,
+        "_terminal_config_bridge_state",
+        {"home": hermes_home_key(), "owned": set()},
+    )
     yield
     terminal_tool._task_env_overrides.clear()
     terminal_tool._task_env_overrides.update(before_overrides)
--- a/tests/tools/test_terminal_degraded_mode.py
+++ b/tests/tools/test_terminal_degraded_mode.py
@@ -27,9 +27,12 @@ def isolated_env(tmp_path, monkeypatch):
     import tools.terminal_tool as tt
 
     monkeypatch.setenv("HERMES_HOME", str(tmp_path / ".hermes"))
-    # The one-shot config bridge would overwrite our TERMINAL_* test vars
-    # from the developer's real config.yaml; mark it as already attempted.
-    monkeypatch.setattr(tt, "_terminal_config_bridge_attempted", True)
+    # The config bridge would overwrite our TERMINAL_* test vars from the
+    # developer's real config.yaml; mark it as already run for this home.
+    from hermes_constants import hermes_home_key
+    monkeypatch.setattr(
+        tt, "_terminal_config_bridge_state", {"home": hermes_home_key(), "owned": set()}
+    )
 
     def _clear():
         with tt._env_lock:
--- a/tests/tools/test_terminal_env_bridge.py
+++ b/tests/tools/test_terminal_env_bridge.py
@@ -11,13 +11,13 @@ import os
 import pytest
 
 import tools.terminal_tool as terminal_tool
-from hermes_constants import get_hermes_home
+from hermes_constants import get_hermes_home, reset_hermes_home_override, set_hermes_home_override
 
 
 @pytest.fixture(autouse=True)
 def _reset_bridge_state(monkeypatch):
     """Each test starts with an un-attempted bridge and clean mapped env."""
-    monkeypatch.setattr(terminal_tool, "_terminal_config_bridge_attempted", False)
+    monkeypatch.setattr(terminal_tool, "_terminal_config_bridge_state", None)
     for name in (
         "TERMINAL_ENV",
         "TERMINAL_CWD",
@@ -152,3 +152,65 @@ def test_bridge_config_failure_does_not_crash(monkeypatch):
 
     assert config["env_type"] == "ssh"
     assert config["ssh_host"] == "example.test"
+
+
+def test_profile_switch_serves_each_profiles_terminal_backend(tmp_path):
+    """A unified dashboard backend serves non-default profiles under a
+    context-local HERMES_HOME override; each profile must execute commands
+    with ITS configured terminal backend, not the launch profile's (#98581 —
+    profiles with ``terminal.backend: docker`` ran on the host instead)."""
+    launch_home = tmp_path / "root"
+    docker_home = tmp_path / "root" / "profiles" / "web"
+    docker_home.mkdir(parents=True)
+    (launch_home / "config.yaml").write_text("terminal:\n  backend: local\n")
+    (docker_home / "config.yaml").write_text(
+        "terminal:\n  backend: docker\n  docker_image: sandbox/image:1\n"
+    )
+
+    launch_token = set_hermes_home_override(str(launch_home))
+    try:
+        assert terminal_tool._get_env_config()["env_type"] == "local"
+
+        profile_token = set_hermes_home_override(str(docker_home))
+        try:
+            config = terminal_tool._get_env_config()
+        finally:
+            reset_hermes_home_override(profile_token)
+
+        assert config["env_type"] == "docker"
+        assert config["docker_image"] == "sandbox/image:1"
+        assert os.environ["TERMINAL_ENV"] == "docker"
+
+        # Back on the launch profile: the profile's docker selection must
+        # not stick, and the launch backend applies again.
+        assert terminal_tool._get_env_config()["env_type"] == "local"
+        assert os.environ["TERMINAL_ENV"] == "local"
+    finally:
+        reset_hermes_home_override(launch_token)
+
+
+def test_profile_switch_drops_launch_terminal_selection(tmp_path):
+    """TERMINAL_* vars owned by the launch profile's bridge must not survive a
+    switch to a profile without a terminal section — the unconfigured profile
+    falls back to the local default, not the launch profile's backend."""
+    launch_home = tmp_path / "root"
+    plain_home = tmp_path / "root" / "profiles" / "plain"
+    plain_home.mkdir(parents=True)
+    (launch_home / "config.yaml").write_text("terminal:\n  backend: docker\n")
+    (plain_home / "config.yaml").write_text("agent:\n  max_turns: 100\n")
+
+    launch_token = set_hermes_home_override(str(launch_home))
+    try:
+        assert terminal_tool._get_env_config()["env_type"] == "docker"
+        assert os.environ["TERMINAL_ENV"] == "docker"
+
+        profile_token = set_hermes_home_override(str(plain_home))
+        try:
+            config = terminal_tool._get_env_config()
+        finally:
+            reset_hermes_home_override(profile_token)
+
+        assert config["env_type"] == "local"
+        assert os.environ["TERMINAL_ENV"] == "local"
+    finally:
+        reset_hermes_home_override(launch_token)
--- a/tools/terminal_tool.py
+++ b/tools/terminal_tool.py
@@ -1703,11 +1703,18 @@ def _is_unusable_container_cwd(cwd: str) -> bool:
     return False
 
 
-# One-shot guard for the config-fallback bridge below.  Purely an
-# optimization: after the first attempt either TERMINAL_ENV is set (bridge
-# succeeded — merged config always carries terminal.backend) or the import
-# failed and retrying every call would be wasted work.
-_terminal_config_bridge_attempted = False
+# Guard for the config-fallback bridge below, keyed by Hermes home.  The
+# bridge runs at most once per active home: within one home it is purely an
+# optimization (after the first attempt either TERMINAL_ENV is set or the
+# import failed and retrying every call would be wasted work), but a unified
+# dashboard/serve process serves several profiles through the context-local
+# HERMES_HOME override, and each profile must bridge its OWN terminal.* config
+# (#98581).  ``owned`` are the TERMINAL_* vars the previous home's bridge (or
+# the launch-time CLI bridge) wrote, so a profile switch can drop them before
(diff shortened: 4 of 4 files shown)


liuhao1024:
Note for reviewers: issue #98581 was flagged by triage as a duplicate of #85084 (context-local terminal config scoping for multiplexed gateway turns). Both address the same underlying class — process-global `TERMINAL_*` state leaking across profiles — but at different layers:

- #85084 reworks the **resolution layer** (contextvars-based terminal config scope routing terminal/file/code-execution/env-hint lookups). It is the broader architectural fix, but has been open since 08-13 with no human review yet.
- This PR is a targeted fix at the **bridge layer**: the one-shot module-global guard in `terminal_tool._ensure_terminal_env_bridged()` is re-keyed per active Hermes home, mirroring the per-profile re-bridge `web_server.py` already performs for the embedded chat PTY. One core file, closes the unified-dashboard sandbox escape from #98581 immediately, and the regression tests pin the local→docker→local round-trip that #85084's suite does not cover.

If maintainers prefer landing #85084's scoping approach, I'm happy to close this in its favor — but as long as that broader PR is unreviewed, this small fix removes a live sandbox-escape path for unified-dashboard deployments.

alt-glitch:

Duplicate of #94206: this is a newer iteration of the same open author's per-profile terminal environment bridge repair. #94206 remains the canonical open PR; please consolidate the improved ownership/keying details there.

andrexibiza (commented):
Reviewed exact head `698faaa8c22dd7488f2b19809670e350e7cfd334` against exact current `main@66666f6e2eca0ae883195a34c66131985ea7dd06`. This is one commit directly on current main (`1 ahead / 0 behind`), and exact-head CI `33314107595`, Docker `33314107260`, and Nix `33314107275` are green; contributor attribution is green as well. There were no prior submitted reviews on this head.

The sequential handoff bug is real, and the new `hermes_home_key()` + explicit `terminal_config_owned_env_vars()` ownership is a cleaner bridge implementation than the old one-shot. I do have one security blocker before this can be treated as closing the sandbox-isolation class.

### P1 — the lock serializes the *rewrite*, not the profile-scoped read, so concurrent turns can still cross backends

`_terminal_config_bridge_lock` only covers `_ensure_terminal_env_bridged()`. The consumers then read the shared process environment **after the lock has been released**. In this exact head, `_get_env_config()` does:

```py
_ensure_terminal_env_bridged()
env_type = os.getenv("TERMINAL_ENV", "local")
mount_docker_cwd = os.getenv("TERMINAL_DOCKER_MOUNT_CWD_TO_WORKSPACE", ...)
```

and `_session_isolation_enabled()` / `_docker_persistent_profile_scoped()` have the same `ensure -> os.getenv(...)` shape.

That leaves a deterministic interleaving in the same unified process:

1. profile A (`docker`) enters `_get_env_config()`, bridges A into `os.environ`, then `_ensure_terminal_env_bridged()` releases the lock;
2. before A reads `TERMINAL_ENV`, profile B (`local`) enters, takes the lock, purges A's owned keys, bridges B, releases;
3. A resumes and reads B's `TERMINAL_ENV=local`.

A turn that is supposed to be Docker-sandboxed can therefore still resolve as host-local. The new tests are all serial (`local -> docker -> local`, and `docker -> profile-without-terminal`), so they prove handoff correctness but cannot witness this overlap.

This is not an invented stronger contract: canonical issue #68559 explicitly requires **concurrent interleaved turns to remain isolated** and says profile-specific terminal policy must not be implemented by mutating `os.environ` per turn. Its proposed invariant is a context-local terminal configuration consulted before process-global defaults. The original implementation lineage is closed-unmerged #68611 by `x7peeps`; open #79117 by `100yenadmin` rematerializes that ContextVar shape and includes the load-bearing `copy_context()` worker-hop regression. Open #85084 by `CannibalKush` is broader/staler, but it reaches the same ownership conclusion and additionally carries terminal/file/code-execution and environment-identity work.

**Required fix/proof:** make the routed profile's resolved terminal policy immutable/context-local for the duration of the turn (with the existing process env only as the single-profile/launch fallback), and add an adversarial overlap regression that pauses A *after* its profile policy is established, lets B establish a different backend, then resumes A and proves A still resolves Docker. Keep the sequential controls too. If the implementation uses a ContextVar, the test also needs the real worker-carrier/copy-context seam so the proof matches the execution boundary rather than only one asyncio task.

### Ownership / merge topology

There are two distinct pieces of prior work that should not be flattened:

- **#94206 (`liuhao1024`) is the direct prior version of this exact bridge-level repair.** #98589 is a materially cleaner/current superseding iteration, not independent work. Preserve that lineage explicitly and keep any stronger predecessor regressions that still apply rather than leaving two competing `tools/terminal_tool.py` owners open.
- **#97014 (`Bergmann89`) is complementary, not duplicate.** It fixes `env_loader._reapply_terminal_config_bridge()` and states the critical process boundary plainly: in a multiplex process, shared `os.environ` belongs to the launch profile; secondary profiles must reach execution through session/context-scoped state, never by rewriting the shared environment. Landing #98589 as written would encode the opposite ownership model inside `terminal_tool`, so these need convergence rather than independent merge order.

The issue graph also needs correction before closure. #98581 is already labeled duplicate and triaged to #85084; #94200 is labeled duplicate and triaged to canonical #68559. `Fixes #98581` currently makes this look like class closure even though the concurrent/no-process-global-mutation half of #68559 remains open on this head. Please interlock the canonical class explicitly and either narrow the closing claim to the proven sequential dashboard handoff or converge on the context-local owner and close the whole shape.

The exact-head CI is excellent and the sequential re-bridge improvement is useful work. The remaining issue is the authority carrier itself: profile-specific sandbox policy cannot safely live in one mutable process-wide environment when two profiles can execute at once.

liuhao1024:
Closing in favor of #94206, per the triage note — both PRs are mine against the same root cause in `_ensure_terminal_env_bridged`, and #94206 (older, already carrying an AI review round) is the right place to keep the history.

The ownership/keying improvements from this PR are now consolidated into #94206 (commit f87ba8d2): home-keyed scope via `hermes_home_key()`, config-section-owned keys via `terminal_config_owned_env_vars` (which also drops launcher-bridged selections the diff-based tracking missed), plus this PR's dashboard regressions. That branch now covers both #94200 and #98581, so nothing is lost by closing this one.

PR #98590, opened 2026-08-30
fix(desktop): honour arrowed slash-picker highlight on bare '/' query (Enter sends correct command)

### LINKED ISSUE #98535: Slash command picker: Enter after arrow-key selection on a bare `/` sends an empty command instead of the highlighted command
## Bug

In the desktop composer, typing a bare `/` opens the slash command picker. **Arrow-key selecting a command and pressing Enter sends an empty command** (backend reports "empty command") instead of executing the highlighted command.

## Repro

1. Open a chat in the desktop app.
2. Type `/` (nothing else). The command picker opens.
3. Press ArrowDown/ArrowUp to highlight a command (e.g. `/help`).
4. Press Enter.

**Expected:** the highlighted command is committed/sent (the picker's own comment says "Enter means 'I want this one'").
**Actual:** the selected item is ignored, a bare `/` is sent, and the backend replies "empty command".

Note: typing a prefix first (e.g. `/hel` + Enter) works fine — only the bare-`/` + arrow-key flow is broken.

## Root cause

`apps/desktop/src/app/chat/composer/composer-utils.ts` — `implicitSlashAcceptIndex()`:

```ts
export function implicitSlashAcceptIndex(query, itemTexts, activeIndex, activeExplicit) {
  const typed = slashCompletionToken(query)   // "" for a bare "/"

  if (!typed) {
    return null        // <-- returns null BEFORE the explicit-pick branch
  }

  if (activeExplicit && itemTexts[activeIndex] != null) {
    return activeIndex // <-- never reached for a bare "/" query
  }
  ...
}
```

The `!typed` guard fires first, so even an explicitly arrowed highlight is refused on a bare `/`. The caller (`composer/index.tsx`) then does `triggerItems[implicitSlashAcceptIndex(...) ?? -1]` → `triggerItems[-1]` → undefined → the accept path is skipped and the keypress falls through to message send, emitting the bare `/`.

This also contradicts the function's own doc comment ("Enter still accepts once the user has arrowed to a row deliberately, so the highlight never lies about what Enter will do") and `acceptsTriggerCompletion()`, which already returns true for Enter when `activeExplicit`.

## Suggested fix

Move the explicit-pick branch above the `!typed` guard:

```ts
export function implicitSlashAcceptIndex(query, itemTexts, activeIndex, activeExplicit) {
  // A deliberately arrowed highlight always wins: Enter means "I want this
  // one", even on a bare `/` query where no command name has been typed yet.
  if (activeExplicit && itemTexts[activeIndex] != null) {
    return activeIndex
  }

  const typed = slashCompletionToken(query)

  if (!typed) {
    return null
  }
  ...
}
```

Suggested regression tests for `composer-utils.test.ts`:

```ts
it('honours an arrowed pick on a bare `/` query', () => {
  expect(implicitSlashAcceptIndex('', rows, 1, true)).toBe(1)
})

it('still refuses a bare `/` query without an explicit pick', () => {
  expect(implicitSlashAcceptIndex('', rows, 0, false)).toBeNull()
})
```

Both the reordered implementation and the new tests pass locally (existing suite: 25/25).

salch-cred:
## Problem (#98535)

In \implicitSlashAcceptIndex()\, the \!typed\ guard fires **before** the \ctiveExplicit\ branch. On a bare \/\ query, \slashCompletionToken('/')\ returns \''\ which is falsy, so the function returns \
ull\ immediately — even when the user arrow-key selected a specific command. The caller then does \	riggerItems[-1]\ (undefined) and falls through to sending the bare \/\.

## Fix

Move the \ctiveExplicit\ guard above the \!typed\ early-return. A deliberately arrowed highlight always wins regardless of query content.

## Tests

Two new regression cases in \composer-utils.test.ts\:
- bare \/\ + arrow-key → returns the highlighted index
- bare \/\ + no arrow-key → still returns \
ull\

Fixes #98535

### DIFF
--- a/apps/desktop/src/app/chat/composer/composer-utils.test.ts
+++ b/apps/desktop/src/app/chat/composer/composer-utils.test.ts
@@ -112,6 +112,18 @@ describe('implicitSlashAcceptIndex', () => {
     expect(implicitSlashAcceptIndex('review', rows, 0, true)).toBe(0)
   })
 
+  // Regression for #98535: bare `/` + arrow-key selection must commit the
+  // highlighted row, not return null.  Before the fix, `!typed` fired first
+  // and the explicit-pick branch was never reached.
+  it('honours an arrowed pick on a bare `/` query', () => {
+    expect(implicitSlashAcceptIndex('/', rows, 1, true)).toBe(1)
+    expect(implicitSlashAcceptIndex('/', rows, 2, true)).toBe(2)
+  })
+
+  it('still returns null on a bare `/` with no arrow-key selection', () => {
+    expect(implicitSlashAcceptIndex('/', rows, 0, false)).toBeNull()
+  })
+
   it('matches an arg-stage prefix against the full completion text', () => {
     expect(implicitSlashAcceptIndex('personality alic', ['/personality alice', '/personality none'], 0, false)).toBe(0)
   })
--- a/apps/desktop/src/app/chat/composer/composer-utils.ts
+++ b/apps/desktop/src/app/chat/composer/composer-utils.ts
@@ -104,14 +104,18 @@ export function implicitSlashAcceptIndex(
   activeIndex: number,
   activeExplicit: boolean
 ): number | null {
+  // A deliberately arrowed highlight ALWAYS wins — "Enter means I want this
+  // one", even on a bare `/` query where no command name has been typed yet.
+  // This must be checked before the `!typed` early-return so an explicit pick
+  // is never suppressed on a bare `/` (#98535).
+  if (activeExplicit && itemTexts[activeIndex] != null) {
+    return activeIndex
+  }
+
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Fixes #98535: bare `/` query with arrowed highlight now commits the highlighted row on <Enter>.

**Correctness:**
- `apps/desktop/src/app/chat/composer/composer-utils.ts:36-38` — moves `if (activeExplicit && itemTexts[activeIndex] != null) return activeIndex` before `if (!typed) return null`. Previously `!typed` early-return suppressed explicit pick.
- Tests `composer-utils.test.ts:12-18` cover arrowed pick on bare `/` and no-pick null case.

**Non-blocking:** None.

**Verdict:** LGTM — minimal, correct ordering fix.

PR #98591, opened 2026-08-30
fix(desktop): bypass alreadyActive cache skip on explicit pathname navigation to fix SSH session click

### LINKED ISSUE #97809: [Bug]: Desktop (SSH mode): clicking an existing session sends no RPC at all — "Couldn't load this session" while gateway is healthy
### Bug Description

Environment

Hermes Desktop v0.20.6, macOS (Mac mini M4)
Hermes Agent v0.20.6 (2026.8.27), upstream aff5125f, Debian 13 LXC on Proxmox
Gateway mode: Connect via SSH (host 10.0.10.101, user root, key auth)
Model: qwen3.6:35b-a3b via remote Ollama

Summary

In SSH gateway mode, opening any existing session in the Desktop client shows the loading animation, then fails with "Couldn't load this session — The connection to this session failed and automatic retries gave up. Check that the gateway is running, then try again."

The gateway is running. The error message is misleading: the client never sends a resume request at all.

Evidence that the backend is healthy

Gateway PID is unchanged before, during and after the click; RSS moves by a few KB — nothing crashes or restarts
sessions.changed events keep arriving over the WebSocket every ~60s during and after the failure, so the connection is alive
No log entry appears anywhere on the click: not in desktop.log on the client, not in gui.log or the gateway journal on the server. No RPC leaves the client
hermes sessions list on the server returns all 20 sessions instantly; SQLite select count(*) from sessions responds without delay
state.db 4.8M / WAL 2.3M, no lock contention
The same session loads fine in the web dashboard over the same SSH tunnel against the same backend (hermes dashboard --port <p> + ssh -L), full transcript, model live, sending messages works

Ruled out: OOM (memory.events shows oom_kill 0), disk space, DB corruption, model backend unreachable, stale WAL, gateway crash.

Note: this looks distinct from #38873 (remote→local backend flap). Here the client stays correctly connected to the remote backend throughout — desktop.log shows a clean Remote Hermes backend is ready and no fallback. The resume request is simply never issued.

Side observation: every reconnect spawns a new remote dashboard process (hermes serve --isolated, random port) without reaping the previous one. After a few days several orphaned processes accumulate, each holding open handles on state.db.

### Steps to Reproduce

1. Connect Desktop to a remote gateway via SSH mode
2. Confirm connection is established (sessions list renders correctly with real titles and message counts)
3. Click any existing session

### Expected Behavior

Expected: session transcript loads.

### Actual Behavior

Actual: loading animation, then the error above. Creating a new session works normally.

### Affected Component

CLI (interactive chat)

N/A (CLI only)

### Debug Report

```shell
Report       https://paste.rs/WypeL
agent.log    https://dpaste.com/A4V9ZZDX4
gateway.log  https://paste.rs/5Cg2o
gui.log      https://paste.rs/ffXFw
desktop.log  https://dpaste.com/2ABP5AKAR
```

### Operating System

Debian 13 LXC on Proxmox

### Python Version

3.11.16

### Hermes Version

v0.20.6

```shell

```

_No response_

### Proposed Fix (optional)

_No response_

salch-cred:
## Problem (#97809)

In SSH/remote mode, a cached runtime id from before a disconnect makes \lreadyActive=true\ even when the session runtime no longer exists on the gateway. The resume gate (\use-route-resume.ts\) skips the RPC because none of \(gatewayBecameOpen | explicitlyRequested | !alreadyActive)\ is true, leaving the user on the loading spinner forever after clicking an existing session.

## Fix

Include \pathnameChanged\ in the bypass condition. An explicit sidebar click always changes the pathname, so the resume RPC always fires regardless of the cached runtime id. The \
esumeSession\ fast-path restores cheaply if the runtime IS still live; if not, a fresh one is obtained.

The \lreadyActive\ optimisation is still honoured on in-place reconnects and gateway-became-open events where \pathnameChanged\ is false.

Fixes #97809

### DIFF
--- a/apps/desktop/src/app/session/hooks/use-route-resume.ts
+++ b/apps/desktop/src/app/session/hooks/use-route-resume.ts
@@ -164,7 +164,14 @@ export function useRouteResume({
       // rebinds/reaps the session on its side, and trusting it strands Desktop on
       // a dead id ("session not found"). An explicit plugin reselect similarly
       // bypasses the warm-id skip when the focused transcript disappeared.
-      if ((gatewayBecameOpen || explicitlyRequested || !alreadyActive) && shouldResume && !creatingSessionRef.current) {
+      //
+      // Also bypass `alreadyActive` on an explicit pathname navigation
+      // (pathnameChanged): a sidebar click in SSH/remote mode can arrive with a
+      // stale runtime id cached from before the previous disconnect, making
+      // `alreadyActive=true` even though the session runtime no longer exists on
+      // the gateway.  Trusting the cache in that case silently skips the RPC and
+      // leaves the user staring at the loading spinner forever (#97809).
+      if ((gatewayBecameOpen || explicitlyRequested || !alreadyActive || pathnameChanged) && shouldResume && !creatingSessionRef.current) {
         if (explicitlyRequested) {
           handledResumeRequestRef.current = sessionResumeRequest.sequence
         }


Enough1122:

**Overall:** Fixes infinite spinner on SSH/remote mode sidebar navigation (#97809) where stale cached runtime id made `alreadyActive=true` skip the `resume` RPC.

**Correctness:**
- `apps/desktop/src/app/session/hooks/use-route-resume.ts:17` — adds `|| pathnameChanged` to `if ((gatewayBecameOpen || explicitlyRequested || !alreadyActive || pathnameChanged) && shouldResume ...)`. Explicit pathname change always bypasses warm-id cache skip, forcing rebind/reap path.

**Non-blocking nit:**
- Ensure `pathnameChanged` is stable (derived from location) and not true on every render, else adds redundant RPC. Likely memoized correctly upstream.

**Verdict:** LGTM.

PR #98592, opened 2026-08-30
fix(kanban): add kanban.ready_dispatch config gate to prevent auto-spawn on human-only lanes

### LINKED ISSUE #98560: Kanban: no per-assignee dispatch exemption for 'ready' — only kanban.review_dispatch gates 'review', so a no-model profile still gets spawned and crashes

salch-cred:
## Problem (#98560)

\kanban.review_dispatch: false\ correctly prevents the dispatcher from spawning \
eview\ tasks. There was no equivalent for \
eady\ tasks. A board operator running a human-only 'Product Owner' lane — a real Hermes profile with no model — had no way to stop the dispatcher from claiming those tasks, crashing on startup (no model configured), and looping back to \
eady\ on the next tick.

## Fix

Add \
eady_dispatch_enabled()\ (default \	rue\, mirrors \
eview_dispatch_enabled\) and gate the \
eady_rows\ query on it. Setting \kanban.ready_dispatch: false\ in \config.yaml\ skips the entire ready loop. Both gates are independently configurable.

Fixes #98560

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -9605,6 +9605,24 @@ def has_spawnable_review(conn: sqlite3.Connection) -> bool:
     return False
 
 
+def ready_dispatch_enabled() -> bool:
+    """Return whether ``ready``-status tasks should be auto-dispatched.
+
+    Default is ``True``.  Operators running boards where the "ready" column is
+    always pulled by human-operated terminals (not autonomous Hermes profiles)
+    can set ``kanban.ready_dispatch: false`` in config.yaml to prevent the
+    dispatcher from spawning those tasks and crashing on a no-model profile
+    (#98560).  Mirrors :func:`review_dispatch_enabled`.
+    """
+    try:
+        from hermes_cli.config import load_config
+        return bool(
+            (load_config() or {}).get("kanban", {}).get("ready_dispatch", True)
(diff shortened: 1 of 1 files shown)


Enough1122:

**Overall:** Mirrors `review_dispatch` gate with `kanban.ready_dispatch` for boards where `ready` is human-pulled (#98560). Prevents spawning tasks that crash on no-model profiles.

**Correctness:**
- `hermes_cli/kanban_db.py:9-24` — `ready_dispatch_enabled()` reads `kanban.ready_dispatch` with `True` default, bool coercion, exception-safe fallback `True` (fail-open to preserve legacy behavior).
- `hermes_cli/kanban_db.py:40-45` — when disabled, `ready_rows = []` skips ready dispatch entirely; review dispatch unaffected. Budget split after remains correct (review rows enumerated).

**Non-blocking nit:**
- When `ready_dispatch` is `False`, `has_spawnable_ready()` still returns stale vs new gate? It has no gate yet — but dispatcher won't spawn anyway, so telemetry asymmetry. Consider gating `has_spawnable_ready` similarly if used for "dispatcher stuck" warning (as #98582 did for exempt assignees).

**Verdict:** LGTM.

PR #72056, opened 2026-07-26
fix(cron): preserve actionable drift recovery alerts

sashmatash:
## What changed

- Keep the unpinned provider/model drift guard before agent construction and inference.
- Preserve explicit pin/update guidance for recurring or otherwise repeatable jobs.
- Tell consumed finite one-shots to create a new future job with an explicit provider and model.
- Deliver those lifecycle-specific actions before the generic 180-character failure-summary truncation.
- Correct the cron documentation so `cronjob action=update` is not presented as a remedy for a consumed finite one-shot.

## Why

A finite one-shot is consumed when its dispatch is claimed, even if the later drift guard blocks inference. The existing update-oriented recovery advice therefore pointed to a job that could no longer be updated. The generic chat failure summarizer also truncated both recreation and update guidance before users could see it.

This is a focused follow-up to #44585.

## How to test

```bash
scripts/run_tests.sh \
  tests/cron/test_cron_provider_pin.py \
  tests/cron/test_scheduler.py \
  tests/cron/test_jobs.py -q
```

Result: **396 passed, 0 failed**.

Additional verification:

- Regression tests were observed failing before the delivery-summary fix and passing afterward.
- Mocked finite and repeatable probes both failed closed before `AIAgent` construction and delivered the correct lifecycle-specific action.
- `git diff --check` passed.
- Independent read-only review returned `APPROVED` with no blockers, suggestions, or nits.

## Platform tested

- macOS 26.5.2
- Python 3.11.14

## Notes

The repository-wide suite and Docusaurus build were not run. The documentation change is a single Markdown line; CI remains the authority for full-repository compatibility.

Issue #44585, opened 2026-06-12
[Bug]: Cron can inherit temporary paid provider state and continue billing during pause/stop containment


twe-cloud:
## Summary

A live cron job inherited a temporary global provider change to `provider=nous` / `model=anthropic/claude-fable-5`, continued making paid inference calls after the operator attempted to stop it, and only stopped once the Nous credential/funds path failed.

This created real spend on a protected root-machine runtime where the operator expected a strict no-spend default.

Related existing issue: #19615 (`Cron jobs should distinguish global vs pinned model configuration`), but this report adds a concrete production incident showing the failure mode is more severe than config ambiguity alone.

## Impact

- Unintended paid inference calls against Nous (`anthropic/claude-fable-5`)
- At least one real affected operator incident with reported credit loss of USD $7.73
- Operator pause/stop actions did not immediately contain spend
- Gateway stop was followed by service restart behavior, reopening cron execution

## Environment

- Host: macOS 26.5.1
- Hermes home: profile default under `~/.hermes`
- Runtime surface: gateway + cron enabled
- Repo: `NousResearch/hermes-agent`
- Observed on: 2026-06-11 around 20:10–20:20 local time

## Expected behavior

1. A cron job should not silently inherit a temporary global provider change for paid inference unless explicitly pinned/approved.
2. Pausing a cron job should make it obvious whether an in-flight run will continue, and ideally provide a hard-stop path for the running execution.
3. Stopping the gateway as an emergency containment action should not allow the same spend-risk surface to revive unexpectedly without the operator explicitly choosing restart behavior.
4. For shared live runtimes, provider/auth mutation should fail closed or loudly warn when enabled cron agent jobs could inherit the change.

## Actual behavior

- Cron job `3590b82a5f25` (`Desk Drift daily conversion sales push`) ran with:
  - `provider=nous`
  - `base_url=https://inference-api.nousresearch.com/v1`
  - `model=anthropic/claude-fable-5`
- It continued making repeated paid calls after pause attempts.
- Gateway stop was followed by service restart behavior, and cron resumed.
- First termination attempt did not fully kill the in-flight job tree.
- Nous calls continued until the credential/funds path failed with 401, after which fallback engaged.

## Evidence from logs

From `~/.hermes/logs/agent.log`:

### Job starts on Nous/Fable 5
- `2026-06-11 20:10:13,258 ... provider=nous base_url=https://inference-api.nousresearch.com/v1 model=anthropic/claude-fable-5`
[...]

PR #98595, opened 2026-08-30
fix(compression): stop slow relays from trapping sessions in failed compaction

### LINKED ISSUE #98567: [Bug]: Auto-compression never succeeds on active sessions via slow custom OpenAI-compatible gateway — 256K probe-down vs 1M compression limit mismatch, cache_read explodes (221M)
## Bug Description

Automatic context compression **never succeeds** on a long-lived active session when the model runs through a slow custom OpenAI-compatible gateway: every attempt exhausts the 600s wait ceiling with the summary model producing no output, the session keeps growing, and every subsequent API call re-uploads the full oversized transcript (cache_read exploded to 221M tokens). Two context-length sources disagree (256K probe-down fallback vs 1M compression limit), and a manual `/compress` also fails with "timed out after 120.0s with no output from the summary model".

## Environment

- Hermes Agent v0.20.6 (2026.8.30), Windows 11, Python 3.11
- Custom OpenAI-compatible provider (self-hosted relay gateway, `sub2api` style) for both main and aux/compression routes — same base_url for both
- Main model `glm-5.3-flash` (1M context per models.dev), gateway intermittently returns `502 Upstream service temporarily unavailable`
- Telegram gateway + desktop app sharing one long-lived DM session

## Reproduction

1. Let a Telegram session grow large over 1–2 days (here: 1,600 messages, ~654K tokens estimated, **94% of which are tool-result messages**, dominated by big `execute_code` / `session_search` outputs).
2. Keep the session active (messages keep arriving) while auto-compression triggers.
3. Observe repeated `context compression started` → 600s ceiling reached → `continuing without compression`; messages count keeps rising between attempts (1586 → 1594 → 1600).

## What happened (real logs, 2026-08-30)

```text
19:40:02 INFO [sid] Preflight compression: ~666,239 tokens >= 524,288 threshold (model glm-5.3-flash, ctx 1,048,576)
19:40:03 INFO agent.auxiliary_client: Auxiliary compression: using sub2api (glm-5.3-flash) at https://…/v1/
19:42:02 … still streaming after 120s (last progress 0.0s ago) — extending wait (ceiling 600s)
19:46:02 … still streaming after 359s (last progress 0.3s ago)
19:50:02 WARNING run_agent: Context compression made no progress … (total wait 600.0s, ceiling 600.0s); continuing without compression
```

Manual `/compress` fails identically: `Context compression timed out after 120.0s with no output from the summary model. No messages were dropped`, then `compression is currently blocked (cooldown:60)`.

Usage ledger for the session shows **792 main calls with cache_read = 221,065,088 tokens** and **145 compression calls burning 2.56M input + 1.86M output tokens** — all of it wasted, zero successful commits.

## Root causes (two distinct mechanisms)

**1. Context-length resolution is inconsistent across paths.**

The gateway's `/models` metadata probe fails, so `agent.model_metadata` logs `probe-down` and falls back to **256,000** tokens (`CONTEXT_PROBE_TIERS[0]`), yet compression telemetry simultaneously reports `main_context_limit: 1048576` (1M) — the preflight threshold of 524,288 derives from 1M while the probe path claims 256K. For a custom OpenAI-compatible provider with no context metadata, there is no way to pin one value that both paths honor (`model.context_length` was set to 1048576 after this; the mismatch only surfaced after hours of failed compressions).

This is related to but distinct from #89500 (catalog gap for GLM-5.3 on zai) — here there is *no* catalog entry at all, and the fallback value disagrees with the value compression actually uses.

**2. Auto-compression on an active session with a slow aux route is a lost cause with no early exit.**

The aux summary model is the same slow gateway as the main model (`auxiliary.compression` unset → inherits main model + base_url). First token takes minutes (or never arrives: intermittent 502). The gateway's session-hygiene path caps how long it holds the user's turn (10s) before proceeding uncompressed; the compressor's own wait then runs to the 600s ceiling, gives up, and the session keeps receiving new messages — so the next attempt starts against an *even larger* context. Attempts are serialized behind a cooldown, so the session grows monotonically and no compression ever commits. Unlike #96953 (fence cancelled by `/stop`/`/restart`), there is no restart here — pure timeout/throughput stall on a live session.

## Expected behavior

- A single source of truth for context length per model+provider; the probe-down fallback and the compression limit must not disagree.
- Auto-compression should fail fast (or skip / notify / suggest `/new`) when the summary route has shown no output for the hold budget, instead of repeatedly burning minutes + 2.5M/1.8M input/output tokens per failed round on an actively-growing session.
- A per-attempt (and per-logical-compression) model-call budget so stalled attempts cannot keep consuming aux calls (related to #97648).

## Impact

Massive redundant transmission (221M cache-read tokens on one session), hours of gateway turn latency while compression "waits", and unbounded session growth because compression never commits. For users on self-hosted OpenAI-compatible relays with slower upstreams this turns compression into a pure cost driver with zero benefit.

fangliquanflq:
## What does this PR do?

Long-lived sessions using a slow custom OpenAI-compatible relay can now stop a compression attempt after the configured no-output budget instead of treating empty SSE keepalives as model progress until the 600-second ceiling. When compression inherits the main model and route, it also reuses the main route's explicit context-length override so feasibility checks and the compressor use the same limit.

### Symptom

A custom relay can emit empty streaming events while producing no summary content. Hermes previously reset the compression inactivity clock on every empty event, so an attempt that should fail after 120 seconds remained active until the 600-second total ceiling. The inherited auxiliary feasibility probe could simultaneously report the generic 256K fallback while the main compressor used an explicit 1M context window.

### Impact

Affected sessions repeatedly spend the full compression ceiling without committing a summary, retain the oversized transcript, and resend it on later turns. Cancelled workers can also continue consuming keepalive streams and occupy compression capacity.

### Bug Cause

**Trigger:** `agent/auxiliary_client.py` / `_ChatStreamAccumulator.feed` when a custom OpenAI-compatible relay emits empty SSE chunks, and `agent/conversation_compression.py` / `check_compression_model_feasibility` when compression inherits the main route.

**Causal chain:**
1. The relay emits an SSE chunk with no content, reasoning, or tool-call delta.
2. The stream accumulator marks that transport keepalive as forward progress, continuously extending the host inactivity deadline.
3. The host waits until the total ceiling, returns the original transcript, and later retries compression against an even larger session.
4. Independently, the feasibility check forwards only `auxiliary.compression.context_length`, so an inherited route drops `model.context_length` and can resolve a contradictory fallback limit.

**Why it is wrong:** Transport liveness is not summary progress, and an inherited auxiliary route is the same model-provider endpoint whose context window the user explicitly pinned.

**Working sibling / contrast:** Content, reasoning, and tool-call deltas remain meaningful progress and continue extending the inactivity budget. Separately configured auxiliary models and endpoints still resolve their own context windows.

**Ruled out:** The failure is not caused by the configured 600-second hard ceiling itself. The existing host timeout and cooldown ladder are reached correctly once empty keepalives stop resetting the inactivity clock.

### Fix

- Count only content, reasoning, or tool-call deltas as streamed auxiliary progress.
- Propagate host fence cancellation into the protected auxiliary call and stop consuming the stream on the next event, preventing cancelled attempts from reaching fallback calls or the larger wire ceiling.
- Reuse `model.context_length` only when the resolved auxiliary model and normalized endpoint exactly match the main runtime; preserve independent resolution for distinct auxiliary routes.

## Related Issue

Fixes #98567

## Type of Change

## Changes Made

- `agent/auxiliary_client.py` - distinguish model output from empty keepalive events and stop cancelled streams.
- `agent/conversation_compression.py` - share the main context override with an inherited compression route and propagate fence cancellation.
- `tests/agent/test_aux_progress_streaming.py` - cover empty keepalives and cancellation-driven stream closure.
- `tests/run_agent/test_compression_feasibility.py` - cover inherited and separately routed context overrides.

## How to Test

1. Run the targeted compression, feasibility, and auxiliary-client suites.
2. Confirm empty keepalive chunks do not tick the progress hook while content and reasoning chunks still do.
3. Confirm an inherited auxiliary route receives the main context override and a distinct route does not.

```bash
scripts/run_tests.sh tests/agent/test_auxiliary_client.py tests/agent/test_aux_progress_streaming.py tests/agent/test_compress_context_progress_timeout.py tests/run_agent/test_compression_feasibility.py -q
```

Result: 230 passed on Windows 11.

### Code

- [x] Relevant documentation and docstrings: N/A - no user-facing configuration changed
- [x] `cli-config.yaml.example`: N/A - no config keys changed
- [x] `CONTRIBUTING.md` or `AGENTS.md`: N/A - no architecture or workflow changed
- [x] Cross-platform impact considered - implementation uses platform-independent Python primitives
- [x] Tool descriptions and schemas: N/A - no model tool behavior changed

### DIFF
--- a/agent/auxiliary_client.py
+++ b/agent/auxiliary_client.py
@@ -9218,7 +9218,13 @@ class _ChatStreamAccumulator:
         self.resp_model = model or ""
 
     def feed(self, chunk: Any) -> None:
-        _notify_aux_progress()
+        # A host timeout revokes the compression fence before detaching the
+        # worker. Stop consuming the provider stream as soon as its next event
+        # arrives so empty keepalives cannot keep a cancelled logical attempt
+        # alive (or trigger fallback calls) until the much larger wire ceiling.
+        cancel_check = _capture_aux_cancel_check()
+        if callable(cancel_check) and _captured_aux_cancel_requested(cancel_check):
+            raise AuxiliaryExplicitCancellation()
         if (
             self._total_ceiling is not None
             and (time.monotonic() - self._started) >= self._total_ceiling
@@ -9249,7 +9255,16 @@ class _ChatStreamAccumulator:
         )
         if reasoning_piece and isinstance(reasoning_piece, str):
             self.reasoning_parts.append(reasoning_piece)
-        for tc in (getattr(delta, "tool_calls", None) or []):
+        tool_call_deltas = getattr(delta, "tool_calls", None) or []
+        # Only model output is forward progress. OpenAI-compatible relays may
+        # emit empty SSE keepalive chunks indefinitely; counting those as
+        # progress resets the host's inactivity timer and turns a 120-second
+        # no-output budget into the full 600-second compression ceiling.
+        # Reasoning and tool-call deltas are genuine generation and continue
+        # to keep slow-but-working models alive.
+        if piece or reasoning_piece or tool_call_deltas:
+            _notify_aux_progress()
+        for tc in tool_call_deltas:
             idx = getattr(tc, "index", 0) or 0
             acc = self.tool_calls_acc.setdefault(
                 idx, {"id": "", "name": "", "arguments": []}
--- a/agent/conversation_compression.py
+++ b/agent/conversation_compression.py
@@ -77,6 +77,7 @@ from agent.model_metadata import (
     estimate_request_tokens_rough,
 )
 from agent.session_activity import ActivityProvenance, normalize_activity_provenance
+from hermes_cli.route_identity import normalize_route_base_url
 
 logger = logging.getLogger(__name__)
 
@@ -1775,9 +1776,9 @@ def check_compression_model_feasibility(agent: Any) -> None:
         same_model = str(aux_model or "").strip() == str(
             getattr(agent, "model", "") or ""
         ).strip()
-        same_route = str(aux_base_url or "").rstrip("/") == str(
-            getattr(agent, "base_url", "") or ""
-        ).rstrip("/")
+        same_route = normalize_route_base_url(
+            aux_base_url
+        ) == normalize_route_base_url(getattr(agent, "base_url", ""))
         if aux_context_override is None and same_model and same_route:
             aux_context_override = getattr(agent, "_config_context_length", None)
 
--- a/agent/conversation_compression.py
+++ b/agent/conversation_compression.py
@@ -1760,11 +1760,32 @@ def check_compression_model_feasibility(agent: Any) -> None:
         _raw_aux_key = getattr(client, "api_key", "")
         aux_api_key = "" if (callable(_raw_aux_key) and not isinstance(_raw_aux_key, str)) else str(_raw_aux_key or "")
 
+        # When auxiliary.compression inherits the exact main runtime, its
+        # context window has the same source of truth as the main request.
+        # Reusing only the auxiliary-specific override here made a configured
+        # model.context_length disappear at the feasibility boundary: the main
+        # compressor used the explicit value while this probe fell back to the
+        # generic custom-endpoint tier. Besides contradictory telemetry, that
+        # could lower the live compression threshold to a value the user had
+        # explicitly overridden. A separately routed aux model must still
+        # resolve independently.
+        aux_context_override = getattr(
+            agent, "_aux_compression_context_length_config", None
+        )
+        same_model = str(aux_model or "").strip() == str(
+            getattr(agent, "model", "") or ""
+        ).strip()
+        same_route = str(aux_base_url or "").rstrip("/") == str(
+            getattr(agent, "base_url", "") or ""
+        ).rstrip("/")
+        if aux_context_override is None and same_model and same_route:
+            aux_context_override = getattr(agent, "_config_context_length", None)
+
         aux_context = get_model_context_length(
             aux_model,
             base_url=aux_base_url,
             api_key=aux_api_key,
-            config_context_length=getattr(agent, "_aux_compression_context_length_config", None),
+            config_context_length=aux_context_override,
             # Each model must be resolved with its own provider so that
             # provider-specific paths (e.g. Bedrock static table, OpenRouter API)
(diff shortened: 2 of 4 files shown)


Enough1122:

**Overall:** Two fixes for `#44585`/`#72056` compression path trapping sessions on slow relays.

**Correctness:**
- `agent/auxiliary_client.py:8-13` — `feed()` checks `_capture_aux_cancel_check()` and raises `AuxiliaryExplicitCancellation` when host revoked fence. Prevents empty keepalives from keeping cancelled attempt alive until 600 s ceiling.
- `agent/auxiliary_client.py:13-22` — `_notify_aux_progress()` now only on genuine generation (`piece or reasoning_piece or tool_call_deltas`), not empty SSE keepalives — correctly prevents resetting host inactivity timer (120 s → 600 s bug).
- `agent/conversation_compression.py:17-30` — when aux compression inherits same model+route as main, reuses `agent._config_context_length` instead of falling back to generic tier. Uses `normalize_route_base_url` for URL comparison — avoids contradictory telemetry / lowered threshold.

**Non-blocking nit:**
- `getattr(client, "api_key", "")` callable guard handles lazy keys — good.

**Verdict:** LGTM — well-commented, targeted.

fangliquanflq:
Thanks for the review. I verified the points against the current PR head (`1a93a2d38f12be1dd718761568f8f9c27eeb08bc`):

- Fence cancellation is checked while consuming each streamed event, so a detached compression attempt stops on the next keepalive instead of reaching the wire ceiling or fallback path.
- Only non-empty content, reasoning, or tool-call deltas refresh the progress hook; empty SSE frames no longer extend the inactivity deadline.
- The main context-length override is reused only when the auxiliary model and normalized route match the main runtime, while distinct auxiliary routes continue resolving independently.
- Callable lazy API keys remain excluded from string coercion for metadata lookup.

No follow-up code change is needed for this informational LGTM. Validation: `scripts/run_tests.sh tests/agent/test_auxiliary_client.py tests/agent/test_aux_progress_streaming.py tests/agent/test_compress_context_progress_timeout.py tests/run_agent/test_compression_feasibility.py -q` — 230 passed, 0 failed.

PR #98596, opened 2026-08-30
fix(desktop): promote primary when setting lastUsed so connection switch actually reconnects backend

### LINKED ISSUE #97742: Desktop: switching connection updates lastUsed but not primary, so backend keeps the old connection
## Summary

When switching the active connection in the desktop sidebar between **"This device" (local)** and a **remote gateway**, the *session/chat view* switches correctly, but the **actual backend connection does not follow**: the selected connection is not promoted to `primary`, so the agent backend the desktop talks to stays on the previously-selected connection. The result is that typing into a "local" session silently routes to the *remote* backend (or vice versa), and the input appears to do nothing.

This looks like a more specific, reproducible instance of #94196 (gateway switch requires "Save and reconnect" to restore local backend), but with a sharper root-cause signal: `primary` and `lastUsed` diverge in the connection registry.

## Environment

- Hermes Desktop, packaged build, Windows 11
- Connection registry v2 (`%APPDATA%\hermes\connections.json`)
- Two connections defined: `local` ("This device") and `remote` ("公司", `http://100.95.219.80:9119`, OAuth)
- Hermes Agent version: 1.102.3

## Steps to Reproduce

1. Configure two connections: `local` (This device) and a `remote` gateway.
2. Start the desktop app. With the `remote` connection as the *current* primary, create/open a chat session. (desktop.log shows `Connecting to remote Hermes backend at http://100.95.219.80:9119`.)
3. In the bottom-left connection switcher, switch to **"This device" (local)**.
4. Send a message in what is now the "local" session.

## Expected Behavior

Switching the connection should promote the chosen connection to `primary` **and** tear down/reconnect the backend to that connection — so a "local" session is actually served by the local backend (`127.0.0.1:62187` in our case), fully independent of the remote gateway.

## Actual Behavior

After switching to "This device", the chat UI shows the local session, but the backend is still the remote one. `connections.json` ends up in this inconsistent state:

```json
{
  "primary": "connection",     // <- still the REMOTE connection
  "lastUsed": "local",         // <- the UI switched here, but primary did not follow
  "connections": [
    { "id": "local", "kind": "local", "label": "This device" },
    { "id": "connection", "kind": "remote", "label": "公司", "url": "http://100.95.219.80:9119", "authMode": "oauth" }
  ]
}
```

Because `primary` still points at the remote gateway, the "local" session's input is sent to the remote `serve` process. If that remote `serve` is non-persistent (started in a foreground terminal) or has a transient API error, the turn never completes and the composer appears frozen / unresponsive. Restarting the desktop app does not help unless the user explicitly re-selects Local in Settings → Gateways (which is what actually rewrites `primary`).

## Root-Cause Signal

The switch path appears to update `lastUsed` / the active session view but does **not** call the equivalent of `setPrimaryConnection(registry, id)` (the function that boot uses to decide which backend to connect to, per `electron-main.mjs` ~line 30659). So `primary` is stale relative to the visible connection. Note there are several in-flight PRs in this area (`#96987` keep resolved primary connection identity, `#83219` select a backend per window, `#96895` keep primary SSH session resumes remote) — this report is meant to give those a concrete local↔remote reproduction and the `primary`/`lastUsed` divergence evidence.

## Suggested Fix

On connection switch, promote the selected connection to `primary` and trigger a backend reconnect (tear down old transport, connect to the new `primary`), matching what "Save and reconnect" does manually. Related: #94196.

## Evidence excerpts

desktop.log (startup always connects to remote because `primary` is `connection`):
```
[boot] Connecting to remote Hermes backend at http://100.95.219.80:9119
[boot] Remote Hermes backend is ready
```

agent.log (message accepted by the local view, but the turn never finishes because backend is still remote):
```
14:31:42 tui prompt accepted ... chars=54
14:27:12 api call failed ... provider=opencode-free ... APIConnectionError summary=Connection error
```

salch-cred:
## Problem (#97742)

\hermes:connections:set-last-used\ IPC handler only called \setLastUsedConnection\, which updates \lastUsed\ but leaves \primary\ unchanged. After switching from remote to local (or vice versa) in the connection switcher, \connections.json\ ended up with \lastUsed: 'local'\ but \primary: 'connection'\ (the old remote). The backend kept routing to the previous gateway.

## Fix

Also set \primary = connId\ in the IPC handler. The guard re-checks \connections.some()\ to stay consistent with \setLastUsedConnection\'s own validation.

Fixes #97742

### DIFF
--- a/apps/desktop/electron/main.ts
+++ b/apps/desktop/electron/main.ts
@@ -14733,7 +14733,15 @@ ipcMain.handle('hermes:connections:set-launch-mode', async (_event, mode) => {
   return { ok: true, registry: sanitizeConnectionsRegistry(registry) }
 })
 ipcMain.handle('hermes:connections:set-last-used', async (_event, id) => {
-  const registry = setLastUsedConnection(readDesktopConnectionsRegistry(), String(id || ''))
+  const connId = String(id || '')
+  let registry = setLastUsedConnection(readDesktopConnectionsRegistry(), connId)
+  // Also promote to primary so the backend reconnects to the chosen connection.
+  // Without this, setLastUsed only updated lastUsed while primary kept pointing
+  // at the old connection, and subsequent backend requests silently routed to
+  // the previous gateway (#97742).
+  if (registry.connections.some(c => c.id === connId)) {
+    registry = { ...registry, primary: connId }
+  }
   writeDesktopConnectionsRegistry(registry)
 
   return { ok: true, registry: sanitizeConnectionsRegistry(registry) }


Enough1122:

**Overall:** Fixes connection switch not taking effect (#97742). Setting `lastUsed` now also promotes to `primary` so backend reconnects to chosen connection.

**Correctness:**
- `apps/desktop/electron/main.ts:8-12` — `connId = String(id || '')`, `setLastUsedConnection` then `if (registry.connections.some(c => c.id === connId)) registry = {...registry, primary: connId}`. Guards against promoting non-existent id.
- `writeDesktopConnectionsRegistry` persists both fields; `sanitizeConnectionsRegistry` on return ensures shape.

**Non-blocking:** None — minimal, correct.

**Verdict:** LGTM.

PR #98600, opened 2026-08-30
fix(desktop): clear latched boot-failure overlay when backend recovers after renderer timeout

### LINKED ISSUE #98486: Desktop (Windows): Timed out connecting to Hermes backend - boot overlay latches, recovers only via full process kill (backend/gateway were healthy)
## Summary

On Windows, the desktop app intermittently shows the boot failure overlay **"Hermes couldn't start — Timed out connecting to Hermes backend"** even though the gateway is already running and reachable, and the desktop-spawned backend becomes ready seconds later. The overlay latches permanently: `Retry` works only sometimes (via a renderer bootstrap reset), closing/reopening the window does **not** help (close-to-tray keeps the wedged process alive), and the only reliable recovery is killing all `Hermes.exe` processes and relaunching.

## Environment

- Windows 11 Pro for Workstations, build 26200
- Hermes Desktop (Electron, repo `win-unpacked` build), v0.20.6, observed on commits `9f90cd4`, `2a598aad1c` and earlier
- Gateway run separately (cold-started after `hermes update`), healthy the whole time

## Behavior

Intermittent (roughly every other app open on a busy machine). When it happens:

1. App opens → boot overlay latches with "Timed out connecting to Hermes backend".
2. At the same time the logs show the gateway healthy: `127.0.0.1:8642 LISTENING`, `gateway_state.json → gateway_state: running`, all platforms connected, and the desktop-spawned `hermes serve` backend finishing its own startup seconds later (`HERMES_BACKEND_READY port=…` → `[boot] Hermes backend is ready. Finalizing desktop startup`) — **after** the overlay already latched. That backend-ready event does not clear the overlay.
3. Clicking `Retry` sometimes recovers: the renderer logs `[bootstrap] reset requested by renderer; clearing latched failure` → `Restarting desktop connection` → a fresh backend spawns and becomes ready → UI renders. This proves backend + gateway were reachable and only the renderer boot state was stuck.
4. Closing the window does not reset anything (`closeToTrayOnWindows` keeps the process alive), so reopening resurfaces the same wedged renderer — it looks permanently broken. Only `taskkill /IM Hermes.exe` + relaunch reliably recovers.

### Log timeline from one incident (desktop.log, today)

```
[boot] Starting Hermes backend via Hermes …
[boot] Waiting for Hermes backend to launch          ← renderer dial timeout expires here (busy machine,
                                                       Defender real-time scan slows fresh python.exe spawn)
… ~40s later, backend still finishing:
HERMES_BACKEND_READY port=6522
[boot] Hermes backend is ready. Finalizing desktop startup   ← but overlay is already latched
… after user clicks Retry:
[bootstrap] reset requested by renderer; clearing latched failure
[boot] Restarting desktop connection
HERMES_BACKEND_READY port=28254
[boot] Hermes backend is ready. Finalizing desktop startup   ← recovers immediately
```

## Analysis

- `use-gateway-boot` performs the initial dial with a fixed timeout; once it fails, the connecting overlay (gated on `gatewayState === 'open'`) latches and — apart from the occasional bootstrap-reset path — there is no re-dial when the backend/gateway later become ready.
- So the failure is not "gateway didn't come up"; it is "renderer gave up before the backend was ready and never listens to the ready event afterwards".
- `closeToTrayOnWindows` amplifies it into "every time I open the app it errors", because the wedged process survives window close.

## Possibly related

- #96280 — main-process port announcement never reaches the spawn pipe (uv venv launcher grandchild stdout). That report is deterministic at the Electron↔backend pipe layer; this report covers the renderer-side boot dial latching after the backend becomes ready, with recovery via renderer bootstrap reset. Same recovery dialog, different failing hop.

## Suggested fixes

1. When the main process emits backend-ready (`HERMES_BACKEND_READY` / `[boot] Hermes backend is ready`) after a failed initial dial, automatically re-run the boot connection instead of waiting for a manual reset.
2. Add bounded retry-with-backoff to the initial boot dial (the reconnect machinery exists for post-boot drops, but the first boot dial appears to be effectively single-shot).
3. Optionally: tray menu entry to reload the renderer, so users don't need Task Manager when the overlay latches.

## Workarounds that work today

- Full quit from tray (or `taskkill /IM Hermes.exe`) then relaunch — boots cleanly whenever the gateway is up.
- `Ctrl+R` reloads the renderer and re-dials (recovers in most cases).

salch-cred:
## Problem (#98486)

On Windows with AV scanning, \hermes serve\ can take 60–90 s to bind its port. The renderer's \getConnection()\ times out at ~45 s and latches the boot-failure overlay. When the backend then announces \HERMES_BACKEND_READY\, \pplyDesktopBootProgress\ receives \progress.running = true\ but preserved \current.error\ because \!current.running\ (already failed). The overlay stayed permanently wedged; only killing all \Hermes.exe\ processes recovered.

## Fix

In \oot.ts\, clear the latched error when the new \progress.running\ is \	rue\ — the backend is actively running and the previous failure is superseded. The guard still prevents a late \error: null\ from clearing a current failure.

Fixes #98486

### DIFF
--- a/apps/desktop/src/store/boot.ts
+++ b/apps/desktop/src/store/boot.ts
@@ -35,7 +35,13 @@ export function applyDesktopBootProgress(progress: DesktopBootProgress) {
 
   // Don't let a late progress event (error: null) clobber a previously-set
   // boot failure — failDesktopBoot is terminal for this boot cycle.
-  const error = progress.error ?? (current.running ? null : current.error)
+  // EXCEPTION: if the new progress.running is true, the backend has recovered
+  // (e.g. on a slow Windows machine with AV scanning, the backend announces
+  // HERMES_BACKEND_READY after the 45s renderer timeout already latched the
+  // error overlay — #98486). A live running=true event resets the latch so
+  // the overlay clears and the desktop reconnects rather than staying
+  // permanently wedged until the user kills the process.
+  const error = progress.error ?? ((!current.running && !progress.running) ? current.error : null)
 
   $desktopBoot.set({
     ...current,


Enough1122:

**Overall:** Fixes latched boot-failure overlay wedging desktop on slow Windows/AV where backend announces `HERMES_BACKEND_READY` after 45 s renderer timeout (#98486).

**Correctness:**
- `apps/desktop/src/store/boot.ts:8-12` — changes `const error = progress.error ?? (current.running ? null : current.error)` to `progress.error ?? ((!current.running && !progress.running) ? current.error : null)`. Exception: new `progress.running=true` clears latched error even when `current.running=false` (previous failure) — exactly the recovery case.
- Late `error: null` with `running=false` no longer clobbers `current.error` unless backend also still not running — preserves terminal failure semantics.

**Non-blocking:** None.

**Verdict:** LGTM.

PR #93350, opened 2026-08-24
fix(computer_use): stop killing healthy embedded cua-driver readiness probes at 2s

chelsealong:
## What does this PR do?

Fixes #93312.

`_EmbeddedCuaDaemon.start()` in `tools/computer_use/cua_backend.py` polls
readiness by spawning `cua-driver status --socket ...` with a hardcoded
`timeout=2.0`. cua-driver 0.21.0 introduced a Computer History peer
attestation path that makes an ordinary, healthy `status` call take
2.0-2.35s on macOS (independently reported at trycua/cua#3347). Hermes
killed each probe just before it returned, retried with the same
too-short deadline, and at the 15s outer deadline shut down the
still-healthy daemon, reporting a misleading `embedded cua-driver
startup timed out` error that only showed daemon stderr (proving the
daemon *was* listening) and discarded the actual client-side timeout.

## Changes Made

- `tools/computer_use/cua_backend.py`: give each readiness probe up to
  5s, capped by the remaining overall startup deadline, instead of a
  hardcoded 2s.
- Preserve the last probe's failure detail (timed out, nonzero exit +
  stderr, or OS error) and append it to the final `RuntimeError` so a
  genuine startup failure explains *why* the probe kept failing instead
  of only reporting daemon stderr.
- `tests/tools/test_computer_use_browser_authorization.py`: add
  `test_daemon_probe_timeout_scales_past_two_seconds` (asserts the probe
  timeout passed to `subprocess.run` is now > 2.0s) and
  `test_daemon_startup_error_names_last_probe_failure` (asserts a
  startup failure's error message names the last probe's timeout).

Scope intentionally left out (per the issue's own "proposed fix" list,
items 4-5): switching to a direct socket readiness check instead of
spawning a subprocess per probe, and extending `hermes computer-use
doctor` (tracked separately in #90695) — both are larger, independent
changes.

## How to Test

1. `git checkout HEAD~1 -- tools/computer_use/cua_backend.py` (revert
   just the fix), then run the two new tests — both fail:
   ```
   assert captured_timeouts[0] > 2.0
   E       assert 2.0 > 2.0

   AssertionError: Regex pattern did not match.
   Expected regex: 'last probe: status probe timed out'
   Actual message: 'embedded cua-driver startup timed out: daemon did not become ready'
   ```
2. `git checkout HEAD -- tools/computer_use/cua_backend.py` to restore
   the fix, then re-run — both pass:
   ```
[...]

PR #98601, opened 2026-08-30
fix(computer-use): stop reading a slow readiness probe as a dead daemon

andredezzy:
## The bug

The embedded cua-driver daemon never starts on macOS. Every attempt spends about nineteen seconds and fails with:

```
computer_use backend unavailable: embedded cua-driver startup timed out: daemon did not become ready
```

The tail is empty, so the message blames the daemon — while the daemon is healthy the whole time. `hermes computer-use doctor` reports every probe green, which makes the failure look like a permissions problem it isn't.

## Root cause

The readiness probe runs `cua-driver status --socket <path>` with a two-second timeout. That command opens the daemon socket and **waits for a reply**, and a cold `open -n -g -a CuaDriver.app` launch takes ~2.5s to bind. The probe expired before the first honest answer could arrive, every single time.

`except (OSError, subprocess.SubprocessError)` catches `TimeoutExpired`, so each expiry became `probe = None` — indistinguishable from "daemon says no" — and nothing was recorded. The loop retried until the 15s budget ran out and reported a silence it had produced itself.

Instrumenting the loop against the real driver shows it plainly:

```
 0.2s  rc=1  Cua Driver daemon is not running
 0.3s  TimeoutExpired
 2.4s  TimeoutExpired
 4.5s  TimeoutExpired
 ...
15.1s  TimeoutExpired      <- budget exhausted
```

Eight swallowed exceptions. Meanwhile the same argv, run by hand with a longer probe timeout, is ready in **2.3s**.

## The fix

Three changes, one cause:

- The probe timeout becomes a named constant at **6s**, comfortably past a cold app launch. A probe must outlast the launch it is waiting on.
- `TimeoutExpired` is caught separately and counted. It is not evidence the daemon is dead; it is the absence of evidence either way.
- When the stderr tail is empty the failure now says **which** silence it hit — probes that never got an answer, or a daemon that came up and said nothing. Those are different bugs, and reporting them identically is what made this one hard to find.

CLI discovery (`cua-driver manifest`) is also resolved once per driver binary. It describes a surface that only changes when the binary does, yet it ran on every daemon start.

## Verification

Against the real driver on macOS 26.6.2 / arm64, cua-driver 0.22.2:

```
before:  FAILED after 19.5s
after:   SUCCESS in 2.4s
```

`computer_use` then returns a live capture through the normal tool path.

Tests drive the real `start()` with the daemon and probe stubbed, including a probe that times out twice before answering — the exact shape of the bug. `tests/computer_use/conftest.py` clears the discovery cache between tests, which several existing tests need because they stub different drivers behind one path.

```
48 passed, 5 skipped
```

`test_cua_no_overlay.py::test_serve_process_disables_overlay_when_policy_requires_it` fails identically on unmodified `origin/main` (verified by checking out `origin/main` and running it there). It is pre-existing and untouched by this PR.

### DIFF
--- a/tests/computer_use/conftest.py
+++ b/tests/computer_use/conftest.py
@@ -1,0 +1,17 @@
+"""Shared fixtures for the computer-use backend tests."""
+
+import pytest
+
+from tools.computer_use import cua_backend
+
+
+@pytest.fixture(autouse=True)
+def _clear_mcp_invocation_cache() -> None:
+    """Isolate the CLI-discovery cache between tests.
+
+    Discovery is cached per driver path because the answer only changes when
+    the binary does. Tests break that assumption on purpose: several stub a
+    different driver behind the same path, so a result cached by one would be
+    served to the next.
+    """
+    cua_backend._reset_mcp_invocation_cache()
--- a/tests/computer_use/test_embedded_daemon_startup_budget.py
+++ b/tests/computer_use/test_embedded_daemon_startup_budget.py
@@ -1,0 +1,161 @@
+"""The embedded daemon's startup budget must belong to the daemon.
+
+`_EmbeddedCuaDaemon.start()` resolves the driver's CLI surface (`cua-driver
+manifest`) before it spawns `serve`, and both were measured against one
+15-second deadline. On a machine where discovery is slow the budget is spent
+before the daemon is launched, and the failure surfaces as "daemon did not
+become ready" with an empty stderr tail — blaming the daemon for time it never
+got.
+"""
+
+import subprocess
+import time
+from typing import Any
+
+import pytest
+
+from tools.computer_use import cua_backend
+
+
+class _FakeProcess:
+    """A spawned daemon that stays alive and never writes to stderr."""
+
+    def __init__(self) -> None:
+        self.stderr = None
+        self.terminated = False
+
+    def poll(self) -> None:
+        return None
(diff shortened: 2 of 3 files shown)


alt-glitch:

Duplicate of #93350: both address healthy embedded cua-driver launches being misclassified by the two-second readiness probe. The earlier PR already includes bounded timeout and diagnostic handling.

andredezzy:
The triage is right — closing this in favour of #93350, which predates it by six days and is the better fix.

Concretely, #93350 does one thing mine does not: it clamps each probe to the remaining startup budget (`probe_timeout = min(_PROBE_TIMEOUT_SECONDS, remaining)`). Mine raises the per-probe timeout without that clamp, so a run of slow probes can overshoot the 15s deadline it is supposed to respect. That is a real defect in my version.

The one thing this PR had that #93350 does not — caching `_resolve_mcp_invocation` per driver binary — is not worth carrying over. It came from my first hypothesis, that CLI discovery was eating the startup budget. When I instrumented the real `start()`, discovery measured **0.1s**. Caching it saves nothing meaningful, and I would rather withdraw it than defend it.

I have posted my independent reproduction on #93350, since it was found on a different machine and a different cua-driver version and may be useful corroboration there.

Thanks @alt-glitch — the duplicate call was correct.

PR #98602, opened 2026-08-30
feat(kanban): per-assignee dispatch exemption for human-owner lanes

### LINKED ISSUE #98560: Kanban: no per-assignee dispatch exemption for 'ready' — only kanban.review_dispatch gates 'review', so a no-model profile still gets spawned and crashes

kokhlo:
## Summary

Adds `kanban.non_dispatchable_assignees` (a list of profile names): assignees in the list are skipped by the dispatcher's ready loop and bucketed into `skipped_nonspawnable` — exactly like an assignee that is not a real profile — so a real, addressable profile (a human-owner lane, a review-only role) can opt out of autonomous dispatch entirely (#98560).

The check runs immediately before the existing `profile_exists` guard in `_dispatch_once_locked`, so `recompute_ready()` promotion is untouched and `review_dispatch=false` keeps protecting the review column as before. The gap today: a real profile with no model configured passes `profile_exists`, the dispatcher spawns a worker that exits immediately, and the block-recurrence breaker files the waiting-on-a-human card as `blocked` — on the reporter's fleet this produced 22 stuck human-owner cards and most of the fleet's crash volume.

Config surface follows `review_dispatch`: a `kanban` map key read through `load_config()`, tolerating a bare string, and defaulting to an empty set (no behavior change for existing configs).

Fixes #98560

### DIFF
--- a/hermes_cli/kanban_db.py
+++ b/hermes_cli/kanban_db.py
@@ -9605,6 +9605,24 @@ def has_spawnable_review(conn: sqlite3.Connection) -> bool:
     return False
 
 
+
+def _non_dispatchable_assignees() -> frozenset[str]:
+    """Assignees configured to never be auto-spawned by the dispatcher.
+
+    Reads ``kanban.non_dispatchable_assignees`` (a list of profile names).
+    Lets a real, addressable profile — e.g. a human-owner lane — opt out
+    of autonomous dispatch entirely, the same way ``review_dispatch``
+    lets a whole status opt out, without waiting for the spawn to fail
+    and the block-recurrence breaker to file the card as blocked.
+    """
+    try:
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #98582: both add `kanban.non_dispatchable_assignees` to keep real human-owner profiles out of autonomous dispatch. The earlier PR also covers config defaults and ready/review health accounting.

kokhlo:
Closing in favor of #98582 — it covers the same `non_dispatchable_assignees` gate plus the config defaults and ready/review health accounting this branch deliberately left out of scope. The regression test here (real-but-exempt profile bucketed as nonspawnable before the profile check) is worth carrying over if the author wants it.

PR #98603, opened 2026-08-30
fix(gateway): honour profile inheritance in the gateway's config reader

andredezzy:
> **Stacked on #98116.** That PR adds `inherit: true`; this one fixes the third reader of the same file so the gateway agrees with it. Review after #98116, or together.

## The bug

A Bot Mode room backed by an inheriting profile answers nothing. The UI shows "thinking…" indefinitely while the log fills with:

```
404 - model: z-ai/glm-5.2 | provider=anthropic model=z-ai/glm-5.2
API call failed after 3 retries
```

The profile's configured model is `claude-opus-5` on Anthropic. Nothing in any config file mentions `glm`.

## Root cause

The gateway reads config through `_load_cfg`, which layers the managed overlay and `${VAR}` expansion onto the raw file but never resolved `inherit: true`. A profile that opts into inheritance keeps only its own overrides on disk, so this reader saw **no model at all**.

`_resolve_model` then took its documented last resort:

```python
except Exception:
    return "z-ai/glm-5.2"
```

…via `get_preferred_silent_default_model()` — deliberately a cheap catalog-labeled model rather than an expensive flagship the user never picked. Sound intent, but it fires here on a profile that *does* have a model, and hands a Z.ai model to an Anthropic endpoint that has never heard of it. Every retry 404s, and the failure never reaches the UI.

This is the **third reader of one config file** to need the same rule. `cli.py`'s `load_cli_config` had the same symptom — a model appearing from nowhere — and was fixed in #98116. `_load_cfg` is the remaining one.

## The fix

`_load_cfg` resolves inheritance in the same place it applies the managed overlay, so the gateway and `hermes_cli.config` agree on what a profile's model is.

Inheritance stays best-effort: if the root config is missing or unreadable the profile's own config still stands, matching `_load_inherited_config`'s documented contract.

## Verification

The RED test reproduced the user-visible failure exactly:

```
AssertionError: assert 'z-ai/glm-5.2' == 'claude-opus-5'
```

Three tests cover the reader (inherited model resolves), precedence (the profile's own keys still win over the root), and the symptom (`_resolve_model` no longer falls through to the silent default).

```
tests/tui_gateway/     695 passed
inheritance suites      15 passed
```

Confirmed on a live desktop room: before the fix, five bots produced 108 `glm` 404s and no reply; after, zero `glm` errors and the room routed and answered normally.

`test_gui_surface_toolsets.py::test_holds_exactly_the_gui_affordances` fails on `origin/main` with these changes absent — a pre-existing change-detector assertion about `apply_layout` in the `desktop_ui` toolset, untouched here.


Enough1122:

**Overall:** Implements `inherit: true` profile config inheritance, layering named profile over root `~/.hermes/config.yaml` leaf-by-leaf with profile winning.

**Correctness:**
- `hermes_cli/config.py` — cache key extended with `root_mtime_ns/root_size`, `managed_scope` handling, root parse skipped with warning if missing/unparseable — does not take profile down.
- `cli.py:14-23` — profile loader via `_deep_merge(inherited, file_config)` after `_normalize_root_model_keys`, `inherit` pop before merge — ensures overriding `model.default` keeps root's `model.base_url` (leaf merge, not replace).
- `cli-config.yaml.example:14-28` — docs clearly state fleet-follows-one-setting use case, invalidation on root edit, no-op in root profile.

**Non-blocking nits:**
- `_deep_merge` must be non-mutating for cached `inherited` — verify implementation copies.
- `inherit: true` strict `is True` check (not truthy) prevents accidental `"true"` string enabling — intentional, document maybe.

**Verdict:** LGTM.

andredezzy:
A concurrency review found that `_load_cfg()` read inheritance through the shared global `_cfg_path` after `_load_cfg_raw()` released its lock. A simultaneous read for another profile could replace that path, so profile A inherited profile B's root model.

Commit `af4f3ef4` captures the active config path once and passes it through the raw read and inheritance resolution. The new barrier-based regression test reproduces `A -> model-B` before the fix and returns the correct model for both profiles after it.

Verified: 19 targeted inheritance tests passed, including the deterministic RED-to-GREEN concurrency case; Ruff passed.

andredezzy:
Resolved the upstream refactor conflicts while retaining the captured config path used by both the raw read and inheritance resolution. The branch remains stacked on #98116.

Verification on `7a94edc258`: 19 inheritance tests passed, including the concurrent-profile regression; another 26 profile-scope, display-config, and compression hot-reload tests passed. Ruff passed. GitHub now reports the branch mergeable.

andredezzy:
Closing this. It is stacked on #98116, which @teknium1 declined: profiles are isolated islands by design, so a change in one can never alter another's behaviour. This PR only carries that inheritance into the gateway's config reader, so the decision applies to it too.

Recording what the branch found, since it outlives the design question. `_load_cfg()` resolved inheritance through the shared global `_cfg_path` after `_load_cfg_raw()` released its lock, so a concurrent read for another profile could swap that path mid-resolution and profile A would inherit profile B's root model. Any future reader that resolves a second file after taking a path from shared state has the same shape, whatever the feature.

I am keeping the feature in my own fork, where a change to the root model reaches every bot profile without editing twelve files. Not proposing it again here. Thanks for the review time, and to @Enough1122 for the cache-key catch.

PR #98604, opened 2026-08-30
feat(gateway): prioritize natural paragraph breaks (\n\n) on message chunking

shali10:
### Summary
When a message exceeds a platform's maximum message limit (such as 4096 code units on Telegram), `BasePlatformAdapter.chunk_message` slices the text into multiple chunks.

Previously, it searched only for single newlines (`\n`) and spaces (` `), which frequently sliced lists, markdown tables, and paragraphs right in the middle even when a clean paragraph break (`\n\n`) was available nearby.

### Changes
- In `gateway/platforms/base.py`, check for paragraph boundary (`\n\n`) first in the headroom region.
- If `split_at < _cp_limit // 3`, fallback to single newline (`\n`), then space (` `).
- Added unit test `test_prefers_paragraph_split_over_single_newline` in `tests/gateway/test_platform_base.py`.

### DIFF
--- a/gateway/platforms/base.py
+++ b/gateway/platforms/base.py
@@ -7540,7 +7540,10 @@ class BasePlatformAdapter(ABC):
             else:
                 _cp_limit = headroom
             region = remaining[:_cp_limit]
-            split_at = region.rfind("\n")
+            # Prefer paragraph break first (\n\n), then single newline, then space
+            split_at = region.rfind("\n\n")
+            if split_at < _cp_limit // 3:
+                split_at = region.rfind("\n")
             if split_at < _cp_limit // 2:
                 split_at = region.rfind(" ")
             if split_at < 1:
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Prefer natural paragraph breaks (`\n\n`) when chunking long Telegram messages, falling back to single newline then space.

**Correctness:**
- `gateway/platforms/base.py:5-8` — `split_at = region.rfind("\n\n")` then `if split_at < _cp_limit // 3: split_at = region.rfind("\n")` — avoids picking a paragraph break too near start (would waste budget). Then `if split_at < _cp_limit // 2: rfind(" ")`. Threshold ordering is deliberate.
- Test `test_prefers_paragraph_split_over_single_newline` sets `max_length = len(p1)+20` so both paragraphs exceed limit but `p1` fits — asserts clean split at `\n\n`.

**Non-blocking nit:**
- When content has no `\n\n`, first `rfind` returns `-1` (< cap/3) so fallback to `\n` works — correct. Edge when `region` is < cap/3 itself, still works.

**Verdict:** LGTM.

PR #98605, opened 2026-08-30
fix(curator): audit nested skill_manage operations

zhuiyueya:
## Summary

Follow-up to #97959 and #97964. This PR intentionally does not duplicate the `absorbed_into` schema advertisement covered by #97964.

- Normalize legacy flat and nested `operations[]` `skill_manage` calls before curator auditing.
- Preserve consolidation declarations from nested delete operations so absorbed skills are reported as consolidated instead of stale-pruned.
- Add regression coverage for nested audit evidence, declaration extraction, end-to-end run reports, and batch delete forwarding.

## Tests

- 94 targeted tests passed across curator reports/classification and skill manager suites.
- `git diff --check`
- Windows footgun check passed for the two production files.

This can be merged alongside or after #97964.

### DIFF
--- a/agent/curator.py
+++ b/agent/curator.py
@@ -636,6 +636,50 @@ def _needle_in_path_component(needle: str, path: str) -> bool:
     return False
 
 
+def _iter_skill_manage_operations(
+    tool_calls: List[Dict[str, Any]],
+):
+    """Yield flat operation dicts from legacy and ``operations[]`` calls.
+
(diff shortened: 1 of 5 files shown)


alt-glitch:

Related to #97964, #98538, and #97959. This branch includes the schema-advertisement fix and additionally normalizes legacy and operations[] audit input; please consolidate the overlapping work.

PR #98169, opened 2026-08-29
fix(desktop): keep profile scope when reusing primary remote backend

dyreckt:
## What does this PR do?

Fixes a profile-scoping defect in the Desktop client. When the registry primary is a shared remote/cloud gateway, `ensureRegistryBackend()` reuses the ambient primary descriptor but drops `sharedRemote: true`. The request router only appends `?profile=<profile>` for `sharedRemote` backends, so Capabilities/Skills requests were sent unscoped and the gateway served the default profile while a named profile was selected.

## Related Issue

Fixes #98168

## Type of Change

- [x] \u{1F41B} Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `apps/desktop/electron/main.ts`: add `sharedRemote: true` to the reused primary-remote descriptor in `ensureRegistryBackend()`.
- `apps/desktop/electron/registry-primary-profile-scope.test.ts`: regression test asserting the scoped URL and the flag on the reuse branch.

## How to Test

1. Configure the registry primary as a shared remote/cloud gateway.
2. Select a non-default profile in Capabilities \u2192 Skills.
3. Confirm the list shows the selected profile's skills (request carries `?profile=<profile>`).

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
- [x] My PR contains **only** changes related to this fix
- [x] I've added tests for my changes
- [x] I've tested on my platform: Windows 10 Desktop (remote shared gateway)

### Documentation & Housekeeping

- [x] Documentation: N/A
- [x] Config keys: N/A
- [x] Architecture/workflows: N/A
- [x] Cross-platform impact considered
- [x] Tool descriptions/schemas: N/A

Issue #98168, opened 2026-08-29
Desktop Capabilities/Skills shows default-profile skills when a named profile is selected on a shared remote


dyreckt:
## Bug Description

When a Desktop client connects to a shared remote/cloud gateway (registry primary) and the user selects a non-default profile in the Capabilities \u2192 Skills selector, the Skills list still shows the default profile's skills instead of the selected profile's.

## Root Cause

Electron main's `ensureRegistryBackend()` reuses the ambient primary backend descriptor for the registry primary (non-local, non-ssh), but the reused descriptor omits `sharedRemote: true`. The request router only appends `?profile=<profile>` for `sharedRemote` backends, so the request goes out unscoped and the gateway serves the default profile.

## Expected vs Actual

- Expected: selecting a profile scopes `/api/skills` with `?profile=<selected>`.
- Actual: the request is sent without the profile query, so the default profile's list is returned and displayed.

## Affected area

`apps/desktop/electron/main.ts` — `ensureRegistryBackend()` primary-descriptor reuse branch.

alt-glitch:
> *This was generated by AI during triage.*

Implementation is open in #98169, which restores selected-profile request scoping for the shared-remote reuse branch.

expat-it:
Confirmed reproduction on Windows 11 with Hermes Agent v0.20.6 (2026.8.27), local commit `4f22543509d1b91dc45bcb369447126c5eb14fb7`, using a registered shared remote as the Desktop primary.

Reproduction:
1. Open **Capabilities → Skills**.
2. Select the named profile `service`; then select `review`.
3. Both views display the default profile's 37 local/learned skills as enabled.

Backend control checks on the same host:
- `hermes -p service skills list --source local --enabled-only` → 0 local skills.
- `hermes -p review skills list --source local --enabled-only` → 0 local skills.
- `hermes -p default skills list --source local --enabled-only` → 37 local skills.

The local `ensureRegistryBackend()` registry-primary reuse branch also lacks `sharedRemote: true`, matching the root cause documented here. PR #98169 is merged, so this is not a request for a duplicate fix; it confirms that the v0.20.6 build is affected and that the profile-scoped CLI state itself remains correct. We will verify the Desktop behavior again after updating to a build containing the merged fix.

PR #98607, opened 2026-08-30
fix(desktop): stamp sharedRemote on the registry-primary reuse descriptor

### LINKED ISSUE #98578: fix(desktop): registry-primary REST requests lose ?profile= — every profile settings page renders the primary profile data
# Title

fix(desktop): registry-primary REST requests lose `?profile=` — every profile's settings page renders the primary profile's data

## Summary

On a desktop whose primary backend is a **registered remote gateway** (connections.json `primary` = a remote `url` entry, `authMode: oauth`), all profile-scoped REST reads from the Settings → Model page (`GET /api/config`, `/api/model/options`, `/api/model/auxiliary`) and the Profiles page arrive at the backend with **`profile=None`**, so every profile chip renders the *primary* (default) profile's configuration. Per-profile data on disk is untouched and correct; the webui (which does not ride this IPC layer) renders the same backend correctly, isolating the defect to the desktop's request routing.

Regression window: **v2026.8.19 (v0.20.5) → v2026.8.27 (v0.20.6)**. Verified by A/B rebuild of the same packaged bundle against the same unchanged backend.

## Environment

- Windows desktop app (packaged `apps/desktop/release/win-unpacked` bundle)
- Backend: `hermes serve --host 0.0.0.0 --port 8643` on a WSL2 host, registered in the desktop's v2 connections registry as a remote connection and promoted to `primary` (this is the recommended thin-client setup post the 2026-08 WSL-migration docs)
- 7 named profiles + default on the backend host; each has its own config.yaml with distinct `model.default` + `fallback_providers`

## Evidence (instrumented backend)

Two one-line `[scope-probe]` warnings were added to `GET /api/config`, `GET /api/model/options`, `GET /api/model/auxiliary` in `hermes_cli/web_server.py` (v0.20.6 backend, untouched between runs). The operator clicked through every profile in Settings while the probes captured the arriving `profile=` query value.

**v2026.8.27 bundle (broken):**

```
117×  profile=None          (28× with explicit_only=True)
 21×  profile='default'      13× 'coder'   10× 'designer'
 19×  profile='evolver'       7× 'debug'    7× 'docs'
  7×  profile='orchestrator'
  0×  profile='architect'    <- the active profile never sends its own scope at all
```

Every profile page arrived scope-less; the backend faithfully answered for its own home (default), and the UI painted default's model/provider and auxiliary rows onto every profile.

**v2026.8.19 bundle (same backend, same session, same clicks):** every Settings request arrives correctly scoped, e.g. `profile='orchestrator'`, `profile='coder'`, … Playwright-style cycle through all chips produces one correctly-scoped triple (`config` / `model/options` / `model/auxiliary`) per profile. With the 0.20.5 bundle, each profile page shows its own model.

## Root-cause analysis

`hermes:api` (electron/main.ts) tags renderer REST calls with the active registry `connectionId` and dispatches via `dispatchRegistryApiRequest` → `pathForRegistryBackendRequest(request.path, requestProfile, connection)`:

```ts
function pathForRegistryBackendRequest(path, profile, backend: RegistryBackendRequestScope) {
  return backend.sharedRemote
    ? pathWithProfileScope(path, profile)
    : translateSelfProfileQuery(path, profile, backend.remoteProfile)
}
```

`?profile=` is appended **only when `connection.sharedRemote === true`**. That flag is set in `connectRegistryBackend()` (fresh dial of a `remote`/`cloud` source) but **not** on the descriptor returned by the v1-primary reuse fast paths inside `ensureRegistryBackend()` (both the SSH `reuseMatchingPrimarySshBackend` branch and the "registry primary resolves to this source id" branch return `{...primaryDescriptor, profile, connectionId}` — the primary descriptor built by `buildRemoteConnection` does not carry `sharedRemote`).

For this user the primary IS the registry source (`connections.json: {"primary": "wsl-gentoo", "connections": [{id: "wsl-gentoo", kind: "remote", url: "http://127.0.0.1:8643", authMode: "oauth"}]}`), so **every** request takes the reuse fast path, gets a descriptor without `sharedRemote`, and goes out unscoped. Suspect commit: `fd031488a` "fix(desktop): prove registry-primary backend ownership electron-side; wait for the primary descriptor before boot restore" (2026-08-26) — introduced in exactly the broken window; 0.20.5 predates it.

## Second, related defect: active-profile pages never scope

Independent of the above: when the pointer profile **is** the app's active profile (e.g. architect selected, then Settings opened), pages are served from the backend's own home via `activateOnCurrentSource` — no `?profile=` is emitted even on 0.20.5. That is correct for a local backend, but wrong when the active profile is only a *label* on a remote primary whose home is `default`: architect's pages silently read the backend home (default) too — the operator observed architect + default mirroring each other's edits while the other five profiles rendered correctly. With a remote primary the active-label path and the self-scope path need to agree (either always send the explicit scope, or the SSH/remote alias translation has to map active-label → backend home).

## Impact

- Silent cross-profile data display (and edit-routing) corruption on every remote-primary (thin-client) install; user-visible as "changing the model on one profile changes them all" / "all profiles show the same model".
- Misdiagnosis risk is high: disk configs and the backend are correct; only the desktop's per-profile pages lie. The webui, which scope-maps correctly, shows the right data.

## Requested fix

Set `sharedRemote: true` (or otherwise preserve `request.profile` scoping) on the descriptors returned by the registry-primary reuse fast path, so `pathForRegistryBackendRequest` scopes the path the same way it does for freshly-dialed remote connections — and separately decide the active-profile-label contract for remote primaries.

## Workaround

Rebuild the packaged bundle from v2026.8.19 (correct scoping restored), or fork-patch `sharedRemote: true` onto the reuse-path descriptor.

liuhao1024:
## What does this PR do?

On a desktop whose primary backend is a registered remote/cloud gateway, every profile-scoped REST read dispatched through `dispatchRegistryApiRequest()` arrived at the backend without `?profile=`, so every profile chip in Settings rendered the primary profile's data (#98578).

Root cause: the registry-primary reuse fast path in `ensureRegistryBackend()` (the `registrySourceOwnsPrimaryBackend()` short-circuit introduced in `fd031488a`) returns `{...primaryDescriptor, profile, connectionId}` — without `sharedRemote`. `pathForRegistryBackendRequest()` only appends `?profile=` when the descriptor carries `sharedRemote: true`; otherwise it takes the self-profile-translation branch, which rewrites an existing self-scope but never adds one. A freshly dialed remote/cloud registry connection stamps `sharedRemote: true` in `connectRegistryBackend()`; the reuse fast path silently dropped that contract, so requests went out unscoped and the backend answered from its own home profile.

Fix: stamp `sharedRemote: true` on the descriptor returned by the primary-owns reuse fast path — one line, same contract as the fresh-dial remote/cloud branch right below it. The SSH reuse branch is intentionally untouched: an SSH backend is isolated per profile (its namespace is `remoteProfile`), so translate-only is the correct semantics there.

## Related Issue

Fixes #98578

## Type of Change

## Changes Made

- `apps/desktop/electron/main.ts`: the `registrySourceOwnsPrimaryBackend()` reuse fast path in `ensureRegistryBackend()` now returns its descriptor with `sharedRemote: true`, so `pathForRegistryBackendRequest()` scopes REST paths with `?profile=` exactly like a freshly dialed remote/cloud registry connection.
- `apps/desktop/electron/registry-primary-scope.test.ts` (new): source assertion that the reuse fast path stamps `sharedRemote`, plus a behavioral check tying the stamped descriptor to `pathForRegistryBackendRequest()` output (scoped vs. the pre-fix unscoped shape).

## How to Test

1. `cd apps/desktop && npx vitest run --project electron electron/registry-primary-scope.test.ts` — Observed result: 2 passed.
2. Wider regression face: `npx vitest run --project electron electron/connection-config.test.ts electron/backend-dial-claim.test.ts electron/connection-registry.test.ts` — Observed result: 218 passed.
3. Red/green check: removing the `sharedRemote: true` stamp makes the new source-assertion test fail (`stamps sharedRemote on the primary-owns reuse fast path`), confirming the test locks the contract. Restoring the stamp turns it green again.
4. Full `--project electron` suite: 1960 passed, 1 pre-existing failure in `managed-ssh-update.test.ts` (POSIX managed launcher, exit 127) that reproduces identically on a clean checkout without this change — environment-related, not introduced here.

### Code

- [x] I've run `pytest tests/ -q` and all tests pass — N/A (TypeScript-only change; ran the desktop vitest electron project instead, results above)
- [x] I've tested on my platform: macOS 15 (arm64); the changed code path is platform-agnostic Electron main-process routing

- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — the fix restores the exact scoping contract already used by the fresh-dial branch, no platform-specific behavior introduced

## Screenshots / Logs

Not applicable — main-process routing change; the issue's instrumented-backend probe (`profile=None` for every Settings request) is the observable, and with the stamp those requests resolve through `pathWithProfileScope()` instead of the translation no-op.

### DIFF
--- a/apps/desktop/electron/main.ts
+++ b/apps/desktop/electron/main.ts
@@ -11151,7 +11151,11 @@ async function ensureRegistryBackend(connectionId, profile, managedUpdateCorrela
       return {
         ...primaryDescriptor,
         profile: profileKey,
-        connectionId: id
+        connectionId: id,
+        // Same contract as the fresh-dial remote/cloud branch below: one host
+        // serves every profile of this source, so REST paths must carry
+        // ?profile= instead of relying on the backend's own home scope.
+        sharedRemote: true
       }
     }
   }
--- a/apps/desktop/electron/registry-primary-scope.test.ts
+++ b/apps/desktop/electron/registry-primary-scope.test.ts
@@ -30,7 +30,10 @@ describe('registry-primary REST scoping (#98578)', () => {
   })
 
   it('the stamped descriptor scopes a profile read the same way a fresh-dial remote does', () => {
+    // Remote/cloud primary descriptors carry no remoteProfile — that is the
+    // isolated-SSH contract — so both shapes below only differ by the stamp.
     const reusedPrimaryDescriptor = {
+      remoteProfile: null,
       baseUrl: 'http://127.0.0.1:8643',
       profile: 'architect',
       connectionId: 'wsl-gentoo',
@@ -43,12 +46,15 @@ describe('registry-primary REST scoping (#98578)', () => {
 
     // Pre-fix shape (no stamp): the translation branch leaves the path
     // unscoped, so the backend serves its home profile's config.
+    const unstampedPrimaryDescriptor = {
+      remoteProfile: null,
+      baseUrl: 'http://127.0.0.1:8643',
+      profile: 'architect',
+      connectionId: 'wsl-gentoo'
+    }
+
     expect(
-      pathForRegistryBackendRequest('/api/config', 'architect', {
-        baseUrl: 'http://127.0.0.1:8643',
-        profile: 'architect',
-        connectionId: 'wsl-gentoo'
-      })
+      pathForRegistryBackendRequest('/api/config', 'architect', unstampedPrimaryDescriptor)
     ).toBe('/api/config')
   })
 })
--- a/apps/desktop/electron/registry-primary-scope.test.ts
+++ b/apps/desktop/electron/registry-primary-scope.test.ts
@@ -1,0 +1,54 @@
+import fs from 'node:fs'
+import path from 'node:path'
+import { fileURLToPath } from 'node:url'
+
+import { describe, expect, it } from 'vitest'
+
+import { pathForRegistryBackendRequest } from './connection-config'
+
+const here = path.dirname(fileURLToPath(import.meta.url))
+const mainSource = fs.readFileSync(path.join(here, 'main.ts'), 'utf8').replace(/\r\n/g, '\n')
+
+describe('registry-primary REST scoping (#98578)', () => {
+  it('stamps sharedRemote on the primary-owns reuse fast path so REST keeps ?profile=', () => {
+    const branchStart = mainSource.indexOf(
+      "if (id === registry.primary && source.kind !== 'local' && source.kind !== 'ssh') {"
+    )
+    expect(branchStart).toBeGreaterThan(-1)
+
+    const body = mainSource.slice(branchStart, branchStart + 700)
+
+    expect(body).toContain('const primaryDescriptor = await ensureBackend(profile)')
+    expect(body).toContain('if (registrySourceOwnsPrimaryBackend(registry, id, primaryDescriptor)) {')
+
+    // The reused primary IS the shared remote gateway (one host, every
+    // profile). Without this stamp the descriptor falls into the
+    // self-profile-translation branch, which never appends ?profile=, and
+    // every registry-scoped Settings/Profiles read answers for the
+    // backend's own home instead of the selected profile (#98578).
+    expect(body).toContain('sharedRemote: true')
+  })
+
+  it('the stamped descriptor scopes a profile read the same way a fresh-dial remote does', () => {
+    const reusedPrimaryDescriptor = {
+      baseUrl: 'http://127.0.0.1:8643',
+      profile: 'architect',
+      connectionId: 'wsl-gentoo',
+      sharedRemote: true
+    }
+
+    expect(
+      pathForRegistryBackendRequest('/api/config', 'architect', reusedPrimaryDescriptor)
+    ).toBe('/api/config?profile=architect')
+
+    // Pre-fix shape (no stamp): the translation branch leaves the path
+    // unscoped, so the backend serves its home profile's config.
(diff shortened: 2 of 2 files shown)


alt-glitch:

Duplicate of #98169. Both patches add sharedRemote: true to the same registry-primary reuse return path and restore profile-query scoping.

liuhao1024:
Closing in favor of #98169 — same one-line root cause (the registry-primary reuse return path drops `sharedRemote: true`), and #98169 landed first. The triage call is correct; keeping both would only split review. (#98578, which this PR linked, is an independent report of the same defect as #98168.)

Two notes that may help #98169's review:

- #98578's reporter attached an instrumented A/B probe (v0.20.5 vs v0.20.6: 117 unscoped `profile=None` requests before the stamp vs none after) — usable as acceptance evidence for the fix.
- The stamp must stay specific to this remote/cloud reuse branch: the SSH fast path (`reuseMatchingPrimarySshBackend`) is deliberately isolated-per-profile (its descriptor carries `remoteProfile` and takes the translate-only branch), so it should remain unstamped — #98169 already scopes it exactly this way.

FYI @dyreckt.

PR #98608, opened 2026-08-30
fix(desktop): keep remote profile settings isolated

fangliquanflq:
## What does this PR do?

Desktop now keeps profile-scoped Settings and Profiles REST requests attached to the selected profile when the registry primary reuses an already-running remote or cloud gateway. Without this fix, the reused descriptor loses its shared-gateway scope marker and the backend silently serves the primary/default profile's configuration for every profile page.

### Symptom

On a desktop whose registry primary is a remote OAuth gateway, requests such as `GET /api/config`, `GET /api/model/options`, and `GET /api/model/auxiliary` can omit `?profile=`. Multiple profile pages then render and edit the primary profile's settings instead of their own.

### Impact

Remote-primary desktop users can see and update the wrong profile configuration even though each profile's on-disk data remains distinct. The UI makes separate profiles appear to share model and auxiliary settings.

### Bug Cause

**Trigger:** `apps/desktop/electron/main.ts:11129` / `ensureRegistryBackend()` registry-primary reuse paths

**Causal chain:**
1. A renderer REST request targets a named profile through a remote or cloud registry primary.
2. `ensureRegistryBackend()` reuses the already-running primary descriptor but returns it without `sharedRemote: true`.
3. `pathForRegistryBackendRequest()` treats the descriptor as backend-scoped and does not append `?profile=`, so the shared gateway answers from its primary/default home.

**Why it is wrong:** Remote and cloud registry sources are shared gateways whose request profile is part of the routing contract, regardless of whether the connection was freshly dialed or reused from the legacy primary path.

**Working sibling / contrast:** Fresh `connectRegistryBackend()` remote/cloud descriptors already carried `sharedRemote: true`, so the same request was scoped correctly when it missed the primary reuse fast path.

**Ruled out:** The backend profile handlers are not dropping the query. Instrumented requests arrive with `profile=None` only on the broken desktop route, while the web UI against the same backend remains correctly scoped.

### Fix

A single pure descriptor decorator now applies registry identity and transport scope to fresh and reused descriptors. Remote/cloud sources always retain shared request scoping, including the active profile label; SSH and local descriptors remain backend-scoped and preserve SSH alias translation.

## Related Issue

N/A

## Type of Change

## Changes Made

- `apps/desktop/electron/connection-config.ts` - centralize registry descriptor identity and shared-remote request scope.
- `apps/desktop/electron/main.ts` - apply the same descriptor contract to both registry-primary reuse paths and fresh registry connections.
- `apps/desktop/electron/connection-config.test.ts` - cover active and cross-profile remote/cloud scoping plus the SSH non-regression.

## How to Test

1. Run the focused routing tests:
   `npx vitest run electron/connection-config.test.ts`
2. Run Desktop type checking:
   `npm run typecheck`
3. Run Desktop lint:
   `npm run lint`

### Code

- [x] I've run the relevant Desktop tests and all tests pass

- [x] Documentation update is N/A
- [x] `cli-config.yaml.example` update is N/A
- [x] `CONTRIBUTING.md` or `AGENTS.md` update is N/A
- [x] I've considered cross-platform impact (Windows and macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility)
- [x] Tool description/schema update is N/A

## Screenshots / Logs

N/A - this is a request-routing fix covered by focused descriptor and URL assertions.

### DIFF
--- a/apps/desktop/electron/connection-config.test.ts
+++ b/apps/desktop/electron/connection-config.test.ts
@@ -51,6 +51,7 @@ import {
   savedProfileSsh,
   tokenPreview,
   translateSelfProfileQuery,
+  withRegistryBackendScope,
   withTransientRetries
 } from './connection-config'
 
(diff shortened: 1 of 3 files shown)


alt-glitch:

Related to #98169 and #98578. This is a broader helper-based implementation of the same registry-primary sharedRemote scoping repair; #98169 is the earlier focused carrier. Please consolidate the competing implementations.

PR #88560, opened 2026-08-17
fix(email): require authserv-id before trusting Authentication-Results

JiuYue0820:
## What does this PR do?

`_verify_sender_authentication()` treated the first `Authentication-Results` header as the receiving server's stamp whenever `authserv_id` was empty. On a self-hosted IMAP box that does not prepend its own stamp, that "first" header is the one the sender wrote. A message with only `Authentication-Results: attacker.self; dmarc=pass` then authenticated, which is the deployment shape GHSA-rxqh-5572-8m77 was meant to close.

The function now refuses every header until an `authserv_id` (config `authserv_id` or `EMAIL_AUTHSERV_ID`) is set. Operators whose mail server does not stamp results can still opt out with `require_authenticated_sender: false` / `EMAIL_TRUST_FROM_HEADER=true`.

Existing tests that passed a real provider stamp without a pin now pass `authserv_id="mx.google.com"` so they still cover the parse/align path.

## Related Issue

Related residual of GHSA-rxqh-5572-8m77 / #52801. #56608 hardens alias parsing and a `header.from` DKIM fallback; it does not refuse an unpinned lone header. Filed as a PR directly.

Fixes #

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `plugins/platforms/email/adapter.py`: empty `authserv_id` → fail closed; pin is now required to accept a header; docstring/comment updated
- `tests/gateway/test_email.py`: existing auth cases pass an explicit pin; new `test_empty_authserv_refuses_forged_only_header`

## How to Test

1. `python -m pytest tests/gateway/test_email.py -q` — 40 passed
2. Call `_verify_sender_authentication` with only `Authentication-Results: attacker.self; dmarc=pass` and `authserv_id=""` → rejected
3. Same header with `authserv_id="mx.ourserver.com"` → rejected (id mismatch); a matching pin + `dmarc=pass` still authenticates

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
- [x] My PR contains **only** changes related to this fix/feature (no unrelated commits)
[...]

PR #98611, opened 2026-08-30
fix(email): default authserv_id to the agent From-domain as documented

### LINKED ISSUE #98599: `authserv_id` defaults to empty, contradicting the comment above it — header-injection defense is silently off by default
### Summary

`_authserv_id` is documented to default to the agent's own From-domain, but the code assigns `""` when neither the config key nor the env var is set. With an empty value the `Authentication-Results` selection loop skips its pinning check entirely and trusts the **first** such header on the message — which is exactly the outcome the comment says the default exists to prevent.

This is a security-relevant default, not a cosmetic doc mismatch: it silently disables a defense described three lines above it.

### Environment

- Email platform adapter with `require_authenticated_sender` at its default (`True`)

### Where

`plugins/platforms/email/adapter.py:588-593`. The comment:

```python
# Optional authserv-id to pin Authentication-Results to the operator's
# own receiving server (defends against an injected header that sorts
# first). Defaults to the From-domain of the agent's own address.
```

The code immediately following:

```python
self._authserv_id = (
    extra.get("authserv_id", "") or _get_secret("EMAIL_AUTHSERV_ID", "")
).strip().lower()
```

There is no fallback to the agent's own address. Unless the operator explicitly sets `authserv_id` or `EMAIL_AUTHSERV_ID`, the value is `""`.

### Effect

In `_verify_sender_authentication` (~line 431) the pinning check is guarded by truthiness:

```python
for raw in headers:
    value = " ".join(str(raw).split())
    if authserv_id:
        # authserv-id is the first token before the first ';'
        serv = value.split(";", 1)[0].strip().lower()
        if not _domains_aligned(serv, authserv_id) and serv != authserv_id.lower():
            continue
    trusted = value
    break
```

With `authserv_id == ""` the branch is skipped and the loop takes the first header unconditionally. The function docstring states the assumption this relies on — "That header is prepended by our own server, so the topmost instance is the one we trust" — which holds for many providers but is precisely the assumption the `authserv_id` option was added to stop relying on.

Providers that stamp multiple `Authentication-Results` headers make this concrete: iCloud writes several, and which one sorts first is not something the operator controls.

### Suggested fix

Either implement the documented default:

```python
self._authserv_id = (
    extra.get("authserv_id", "")
    or _get_secret("EMAIL_AUTHSERV_ID", "")
    or _domain_of(self.email_address)      # documented fallback
).strip().lower()
```

...or, if trusting the topmost header really is the intended default, correct the comment so operators are not led to believe a protection is active when it is not. The first option seems closer to the stated intent.

A warning at startup when `require_authenticated_sender` is on and no `authserv_id` resolved would also make the weaker posture visible.

### Note

This is *not* a report that the adapter "only reads the first `Authentication-Results` header." Pinning via `authserv_id` works correctly when it is set — we set `EMAIL_AUTHSERV_ID` and the behavior is right. The defect is narrower: the **default** does not match its documentation, so operators who never set it lose the defense without being told.

kokhlo:
## Summary

The comment above `_authserv_id` promises a default to the From-domain of the agent's own address, but the code leaves the value empty when neither `authserv_id` config nor `EMAIL_AUTHSERV_ID` is set. With an empty value, the pinning branch in `_verify_sender_authentication` is truthiness-guarded off and the loop trusts the **first** `Authentication-Results` header unconditionally — exactly the outcome the documented default exists to prevent (#98599).

This makes the code match its comment: when no explicit value is configured, `authserv_id` derives from the domain part of the agent's own address, so the pinning check engages with zero operator action. Operators who already set `authserv_id` / `EMAIL_AUTHSERV_ID` are unaffected; an address without `@` (misconfigured) degrades to the previous empty behavior.

Two regression tests: the domain default (adapter with `hermes@test.com` → `test.com`) and explicit-env override wins over the default.

Fixes #98599

### DIFF
--- a/plugins/platforms/email/adapter.py
+++ b/plugins/platforms/email/adapter.py
@@ -587,9 +587,16 @@ class EmailAdapter(BasePlatformAdapter):
 
         # Optional authserv-id to pin Authentication-Results to the operator's
         # own receiving server (defends against an injected header that sorts
-        # first). Defaults to the From-domain of the agent's own address.
-        self._authserv_id = (
+        # first). Defaults to the From-domain of the agent's own address —
+        # the fallback the comment always promised: with an empty value the
+        # pinning loop in _verify_sender_authentication skips its check and
+        # trusts the first header on the message unconditionally (#98599).
+        authserv_raw = (
             extra.get("authserv_id", "") or _get_secret("EMAIL_AUTHSERV_ID", "")
+        ).strip()
+        if not authserv_raw:
+            authserv_raw = self._address.rsplit("@", 1)[-1] if "@" in self._address else ""
+        self._authserv_id = authserv_raw.lower()
         ).strip().lower()
 
         # Track message IDs we've already processed to avoid duplicates
--- a/plugins/platforms/email/adapter.py
+++ b/plugins/platforms/email/adapter.py
@@ -597,7 +597,6 @@ class EmailAdapter(BasePlatformAdapter):
         if not authserv_raw:
             authserv_raw = self._address.rsplit("@", 1)[-1] if "@" in self._address else ""
         self._authserv_id = authserv_raw.lower()
-        ).strip().lower()
 
         # Track message IDs we've already processed to avoid duplicates
         self._seen_uids: set = set()
--- a/tests/gateway/test_email.py
+++ b/tests/gateway/test_email.py
@@ -149,6 +149,31 @@ class TestDispatchMessage(unittest.TestCase):
             adapter = EmailAdapter(PlatformConfig(enabled=True))
         return adapter
 
+    def test_authserv_id_defaults_to_own_from_domain(self):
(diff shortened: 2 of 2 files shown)


alt-glitch:

Related to #88560 and #98599. This branch defaults the pin to the agent From-domain; #88560 instead requires an exact receiving-MTA pin. The submitted diff also has an extra closing parenthesis after the self._authserv_id assignment, so these approaches need consolidation before review.

kokhlo:
The leftover closing line is gone as of f895f1a (syntax re-verified against the branch blob — the earlier commits did leave a stale fragment, good catch). On the consolidation question: this branch intentionally keeps the From-domain default because it is what the config comment already documents, while #88560 requires an explicit MTA pin. If maintainers prefer the stricter exact-pin route, happy to rebase onto whatever lands first — the regression tests here hold either way.

PR #96809, opened 2026-08-28
fix(feishu): release adapter resources even when disconnect is cancelled mid-wait

liuhao1024:
## What does this PR do?

Guarantees Feishu adapter resources are released even when `disconnect()` is cancelled mid-wait, which is what a planned gateway restart does.

The gateway-level teardown cancels `FeishuAdapter.disconnect()` after 5s, but the waits inside can legitimately spend 15s (5s WebSocket CLOSE acknowledgement + 10s thread exit). When the cancellation landed on one of those waits, the local cleanup block at the end of `disconnect()` — SDK executor shutdown, seen-id persistence, and the **Feishu app-scoped lock release** — was skipped entirely. The old gateway process stayed alive owning the app-scoped lock, the replacement process repeatedly failed to claim the Feishu app, and the planned restart stalled for minutes (#96801).

The fix extracts the local cleanup into `_local_teardown()` and runs it from `disconnect()`'s `finally`. Everything in it is synchronous and fast, so a single cancellation cannot interrupt it mid-way. The CLOSE/thread-exit waits keep their existing timeouts and warnings; the normal (uncancelled) path is byte-equivalent apart from the refactor.

## Related Issue

Fixes #96801

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `plugins/platforms/feishu/adapter.py` — wrapped `disconnect()`'s wait sequence in `try`, moved the trailing local cleanup (ws refs, event handler, SDK executor, seen-id persistence, app-lock release, disconnected marker) into a new `_local_teardown()` invoked from `finally`.
- `tests/gateway/test_feishu.py` — regression `test_disconnect_local_teardown_survives_cancellation`: a hanging CLOSE coroutine parks `disconnect()` on the 5s wait, the test cancels the task mid-wait, and asserts the app-lock identity is cleared, the disconnected marker ran, and the SDK executor was shut down.

## How to Test

1. `python -m pytest tests/gateway/test_feishu.py -q` — should pass apart from `test_download_remote_document_blocks_connect_time_rebind`, which fails identically on clean `upstream/main` (pre-existing, SSRF-assertion domain, no overlap with this diff).
2. Red/green verified locally: with the pre-fix `disconnect()` restored, the new regression fails (cancellation skips the app-lock release); with this change it passes, and the existing `test_disconnect_sends_websocket_close_frame` still passes.
[...]

PR #98613, opened 2026-08-30
fix(telegram): add CONTACT filter handler to avoid dropping contact cards

### LINKED ISSUE #98598: Telegram shared contacts are dropped silently — no `filters.CONTACT` handler exists
### Summary

When a user shares a contact card in Telegram, the update carries a structured `contact` payload and **no `text`**. The adapter registers handlers for text, location/venue, and media, but none for `filters.CONTACT`, so python-telegram-bot finds no match and drops the update. The agent is never told anything arrived.

### Environment

- Telegram platform adapter, python-telegram-bot

### Where

`plugins/platforms/telegram/adapter.py`. The handler registration block around line 4353 covers location:

```python
app.add_handler(TelegramMessageHandler(
    filters.LOCATION | getattr(filters, "VENUE", filters.LOCATION),
    self._handle_location_message
))
```

There is no `filters.CONTACT` equivalent. For confirmation: the upstream file is 11,088 lines and contains **zero** occurrences of the string `contact`, in any case.

The failure is not just silent, it is **actively misleading**. Because the agent receives no update at all, it cannot distinguish these two situations:

1. the contact was never delivered to it (what actually happens), and
2. it received the contact but lacks permission to do anything with it.

In practice the model reaches for the second explanation, because that is the plausible-sounding one. Ours confidently told the user it did not have permission to access contacts and suggested granting Contacts access in macOS System Settings — a fix for a problem that did not exist, for a message it had never received. The user spent time granting permissions that changed nothing. A dropped update that produces a confident, wrong, actionable explanation is considerably more expensive than a crash.

### Reproduction

1. Connect the Telegram platform.
2. From the Telegram client, share any contact card with the bot (attachment → Contact).
3. Nothing reaches the agent; no handler fires; no log line is emitted.
4. Send a text message immediately after — it is handled normally, confirming the connection is healthy.

### Suggested fix

Register a `filters.CONTACT` handler that renders the structured payload into text on the normal path, mirroring how `_handle_location_message` already handles the analogous case. The Telegram `Contact` object provides `first_name`, `last_name`, `phone_number`, and often a full `vcard`.

At minimum, even a catch-all that logs "received an unhandled update type" would convert this from an invisible failure into a diagnosable one.

### Workaround

We register a local `filters.CONTACT` handler that renders the contact and writes any `vcard` verbatim to disk, so the agent can import the file rather than retyping fields — retyping being where details get invented.

kokhlo:
Fixes #98598.

When a user shares a Telegram contact card, the update carries a structured `contact` payload and no `text`. The adapter registered handlers for TEXT, COMMAND, LOCATION/VENUE, and media, but never for `filters.CONTACT`, so python-telegram-bot dropped the update silently. This change adds a `filters.CONTACT` handler that renders the contact (first_name, last_name, phone_number, vcard) into a text block on the normal message path, mirroring `_handle_location_message`.

### DIFF
--- a/plugins/platforms/telegram/adapter.py
+++ b/plugins/platforms/telegram/adapter.py
@@ -4379,6 +4379,10 @@ class TelegramAdapter(BasePlatformAdapter):
             filters.PHOTO | filters.VIDEO | filters.AUDIO | filters.VOICE | filters.Document.ALL | filters.Sticker.ALL,
             self._handle_media_message
         ))
+        app.add_handler(TelegramMessageHandler(
+            filters.CONTACT,
+            self._handle_contact_message
+        ))
         # Handle inline keyboard button callbacks (update prompts)
         app.add_handler(CallbackQueryHandler(self._handle_callback_query))
         # Inline command picker (@botname <query>) — searchable, uncapped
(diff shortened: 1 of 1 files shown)


Enough1122:

**Overall:** Adds missing `filters.CONTACT` handler so Telegram contact cards are not dropped (#96809 area).

**Correctness:**
- `plugins/platforms/telegram/adapter.py:6-9` — registers `TelegramMessageHandler(filters.CONTACT, self._handle_contact_message)` alongside existing `PHOTO/VIDEO/...` handlers.
- `_handle_contact_message` mirrors `_handle_location_message` pattern: `_effective_update_message`, `_is_user_authorized_from_message` (auth gate), `_should_process_message` + `_should_observe_unmentioned_group_message`, extracts `first_name/last_name/phone_number/vcard`, builds `MessageType.CONTACT` event, applies group observe attribution, `await self.handle_message(event)`. Text prefix `[The user shared a contact.]` + fields gives model usable context.

**Non-blocking nit:**
- Phone number may be sensitive PII — ensure downstream logging redacts or respects `TELEGRAM_HOME_CHANNEL` privacy. Current path goes through normal message pipeline which already handles it.

**Verdict:** LGTM.

PR #80407, opened 2026-08-06
fix(search): stop the Windows path translator from rewriting search regexes

Drexuxux:
## What

`_escape_shell_arg` runs `_bash_safe_path` on whatever it is handed. On Windows that rewrites every backslash to `/`. For a path that is exactly right — bash eats backslashes and MSYS mangles `C:\Users\x` into the `Directory \drivers\etc does not exist` failure class — and it is what the docstring describes.

The content-search builders hand it the regex and the `--glob`/`--include` filter as well:

```python
cmd_parts.append(self._escape_shell_arg(pattern))
cmd_parts.append(self._escape_shell_arg(path))
```

So on Windows the pattern gets the path treatment. Capturing what actually reaches the shell:

```
grep -rnHE --exclude-dir='.*' '/d+' '/c/Users/.../project' | head -n 20
```

`\d+` arrives as `/d+`, `\bword\b` as `/bword/b`, `foo\.bar` as `foo/.bar`. Nothing matches `/d+`, so the search returns zero results and no error — the agent reads a clean "no matches" and moves on, which is worse than a failure it could react to. Every regex escape is affected, on both engines, in `search`, one of the most-used tools in the loop.

## Root cause

f2fcf89c1 ("fix(windows): bash-safe snapshot paths after #63113") swapped the helper's translator:

```diff
-        from tools.environments.local import _windows_to_msys_path
-        arg = _windows_to_msys_path(arg)
+        from tools.environments.local import _bash_safe_path
+        arg = _bash_safe_path(arg)
```

`_windows_to_msys_path` only rewrites drive-qualified paths, so a bare `\d+` passed through untouched. `_bash_safe_path` adds an unconditional `if "\\" in path: path = path.replace("\\", "/")` for mixed MSYS leftovers — correct for its stated job, but it made the helper act on any string containing a backslash, and the helper was already being used for arguments that are not paths. Both commits that built this helper describe it purely in terms of paths (d1ad9a0f5, f2fcf89c1); nothing intended it to reach a regex.

## The fix

Separate the quoting from the translation. `_escape_shell_literal` quotes for the shell without touching separators; the pattern and glob arguments use it. `_escape_shell_arg` keeps its behaviour and is now a thin wrapper — path arguments are unchanged, including the search root:

```
grep -rnHE --exclude-dir='.*' '\d+' '/c/Users/.../project'
```

## Tests and results

[...]

PR #92267, opened 2026-08-22
fix(tools): preserve search regexes on Windows

ogrrd:
## What does this PR do?

Fixes `search_files` regex arguments being rewritten as Windows paths before they reach `rg` or the `grep` fallback.

On native Windows, `_escape_shell_arg()` sends every value through `_bash_safe_path()`. That is correct for filesystem paths, but `_bash_safe_path()` also replaces backslashes with `/`. Regexes such as `AddDataFilePath\("Digital` therefore reached ripgrep as `AddDataFilePath/("Digital` and failed with an unclosed-group parse error.

### Fix

- Separate arbitrary shell quoting from filesystem path normalization, including globs and filename patterns.
- Preserve upstream's bounded native POSIX ripgrep fast path, passing patterns directly as argv without rewriting.
- On shell backends, pass single-line patterns through `stdin_data` to `rg -f -` / `grep -f -`, avoiding shell/path rewriting and Windows argv encoding.
- Preserve actual CR/LF in ripgrep patterns without converting them to regex escapes: send base64 data through stdin, decode after command preprocessing, and expand one quoted argument. A trailing sentinel prevents command substitution from stripping final newlines. This preserves `(?x)` whitespace/comment semantics and fixed-string zero-match probes. CR/LF patterns containing NUL are rejected explicitly; normal OS argument-size limits apply to this multiline path.
- Keep the grep fallback line-oriented rather than turning control characters into false matches.
- Group stdin-consuming pipelines so SDK heredoc redirection feeds the search engine rather than the trailing `head` process. Avoid an accidental blank second pattern.
- Distinguish multiline regex compilation errors from zero matches while retaining usable partial results for ordinary I/O errors.
- Keep the existing MSYS/native handling for actual filesystem paths.

## Related work

This is distinct from #80314, which addressed the search-root path passed to native `rg.exe`. This follow-up addresses regex/glob values accidentally passing through the path normalizer.

## Type of change

- [x] Bug fix (non-breaking change that fixes incorrect behavior)

### Current head: macOS rebase verification (September 21, 2026)

Rebased onto upstream `main` at `3be255eca6552e4f787bdfb0829fe83a1ab810bd`; current head is `0aa705109f1b32b8326ab07d311e1ca6aa6b2e85`. Resolved the filename-search conflict by retaining upstream's `find -H` symlink-root behavior together with this PR's non-path-translating pattern quoting. Upstream hidden-directory handling, live-cwd errors, and bounded native POSIX search remain intact.

[...]

PR #98617, opened 2026-08-30
fix(tools): search patterns are not paths — stop MSYS-translating regex on Windows

Dolverin:
## Behavior
- Preserve opaque regex and glob literals when invoking search tools through Git Bash on Windows.
- Convert filesystem roots separately to native Windows paths; cover matching and no-match paths.
- Keep native Windows tests on the Windows host rather than mocking the OS discriminator.

## Validation
- Official native Windows target suite: **65 passed, 0 failed, 6 Linux-only skipped**.
- Real `LocalEnvironment -> Git Bash -> rg.exe` regression covers backslash regex/glob handling, native roots and no-match behavior.
- Ruff and `git diff --check` passed; exact three-file candidate received independent final review.

## Publication
Updated fork revision: `8821d67c0821c0fbc445c9b1fbfa748e0054b17d`, one reviewed commit on upstream base `91adf584a4783d600e8f358a4f32ad95e7e0cdde`. Hosted CI must be evaluated for this exact revision; previous checks do not validate it. No installation or runtime change is included.

### DIFF
--- a/tests/tools/test_file_operations.py
+++ b/tests/tools/test_file_operations.py
@@ -898,3 +898,97 @@ class TestEscapeNativeToolArg:
         assert node_cmds, f"no node command captured in: {commands}"
         assert "'C:/Users/alice/app/main.js'" in node_cmds[0]
         assert "/c/Users" not in node_cmds[0]
+
+
+class TestSearchPatternIsNotAPath:
+    r"""Regression: a regex pattern must never go through the path translator.
+
+    Live failure (Windows, Aug 2026): `_search_with_rg` / `_search_with_grep`
+    escaped the PATTERN with `_escape_shell_arg`, which calls `_bash_safe_path`
+    — a *path* translator that rewrites every backslash to a forward slash on
+    Windows. Every regex escape was destroyed on the way to the binary:
+
+        setInterval\(   -> setInterval/(    -> rg: unclosed group
+        session\.resume -> session/.resume  -> matches the wrong thing
+        \bfoo\b        -> /bfoo/b          -> matches nothing
+
+    189 searches died this way in one week. A pattern's backslashes are regex
+    escapes, not path separators, so they must survive untouched.
+    """
+
+    def _ops(self, mock_env):
+        return ShellFileOperations(mock_env)
+
+    def _capture(self, mock_env):
+        commands = []
+
+        def side_effect(command, **kwargs):
+            commands.append(command)
+            if "test -e" in command:
+                return {"output": "exists", "returncode": 0}
+            if "command -v" in command:
+                return {"output": "yes", "returncode": 0}
+            return {"output": "", "returncode": 0}
+
+        mock_env.execute.side_effect = side_effect
+        return commands
+
+    @pytest.mark.parametrize(
+        "pattern",
+        [r"setInterval\(", r"session\.resume", r"\bfoo\b", r"\d+", r"toBe\(.*\)"],
+    )
+    def test_escape_shell_literal_preserves_regex_escapes(
+        self, mock_env, monkeypatch, pattern
+    ):
+        import tools.environments.local as local_mod
+
+        monkeypatch.setattr(local_mod, "_IS_WINDOWS", True)
+        ops = self._ops(mock_env)
+        assert ops._escape_shell_literal(pattern) == "'" + pattern + "'"
+
+    def test_escape_shell_literal_still_quotes_single_quotes(self, mock_env):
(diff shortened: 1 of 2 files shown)


alt-glitch:

Duplicate of #80407. Both patches separate regex/glob literal quoting from Windows path translation across the same search paths; #80407 is the earlier open implementation.

Dolverin:
Thanks — the triage is right about the root cause, and #80407 got there first
(2026-08-06). It independently arrived at the same split and even the same
helper name, `_escape_shell_literal`. Credit where it is due.

Two things kept me from just closing this one:

**#80407 no longer merges.** It is `CONFLICTING` / `DIRTY` against current
`main` and has not been touched since 2026-08-06. It was written before
`_escape_native_tool_arg` landed, so it still routes the *path* argument
through `_escape_shell_arg` — which is exactly where it now collides.

**It covered something this PR did not.** #80407 also fixed the `file_glob`
sites; my first commit only covered the pattern. I have now pulled that half
in (`b792401`), so this branch is a superset:

| | #80407 | this PR |
|---|---|---|
| pattern in `_zero_match_probe` | yes | yes |
| pattern in rg / grep / content | yes | yes |
| `file_glob` (`--glob`, `--include`, `-name`) | yes | yes (added) |
| path argument | `_escape_shell_arg` (pre-`_escape_native_tool_arg`) | `_escape_native_tool_arg` |
| tests | 7 | 26 |
| mergeable | no | yes |

The extra tests pin the boundary that makes this bug easy to reintroduce: in a
single command the glob and pattern must stay literal while the path still gets
the native `C:/` form. Without that, a future refactor collapses the two helpers
back together and the bug returns.

Verified on Windows, `pytest tests/tools/test_file_operations.py`:

| | passed | failed |
|---|---|---|
| base `66666f6e2` | 60 | 8 |
| this branch | 80 | 8 |

Identical 8 failures before and after — pre-existing host limitations
(`SeCreateSymbolicLinkPrivilege`, POSIX `umask`). No new failures, +20 passing.

Happy either way: land this one and close #80407, or I can rebase #80407's
authorship forward instead if the maintainers prefer that. The important thing
is that one of them lands — every search pattern carrying a regex escape is
silently broken on Windows today.

monerostar:

Host: Windows 11 (build 26200.9168), Python 3.11.15, pytest 9.1.1, `MSYS_NO_PATHCONV=1`. Checked out `b7924018cb`.

### Live rewrite (this machine)

`_bash_safe_path` / `_escape_shell_arg` really do destroy regex escapes on native Windows:

```
setInterval\(   ->  setInterval/(
session\.resume ->  session/.resume
\bfoo\b         ->  /bfoo/b
\d+             ->  /d+
```

`_escape_shell_literal` keeps the original pattern. Path handling stays split correctly:

- `_escape_shell_arg(r"C:\Users\Admin\src\hermes-agent-contrib")` → `'/c/Users/Admin/src/hermes-agent-contrib'`
- `_escape_native_tool_arg(...)` → `'C:/Users/Admin/src/hermes-agent-contrib'`

That second form is what `rg` needs here (Hermes bash has `MSYS_NO_PATHCONV=1`, so `/c/...` is not converted back).

### Tests

```
python -m pytest tests/tools/test_file_operations.py \
  -k "SearchPatternIsNotAPath or EscapeNativeToolArg or escape_shell" \
  -v --tb=short -o addopts=
```

**28 passed, 60 deselected in 2.67s** — all `TestSearchPatternIsNotAPath` cases plus the native-path tests.

Did not have `rg` on the git-bash PATH for a live binary probe; the construction tests + live `_bash_safe_path` output are the evidence.

### Sibling note (not a rubber-stamp)

- **#80407** is the earlier same-idea patch (`_escape_shell_literal`). Fair duplicate historically. This PR is the better current landing candidate: it is based on later main and keeps `_escape_native_tool_arg` for the search *root*, which #80407 still sends through `_escape_shell_arg`. Landing 80407 as-is on today's tree would risk feeding `/c/...` to native `rg`.
- **#92267** is a third approach (`rg -f -` / stdin). Same bug class; different shape.

The helper split here is small, matches existing quoting, and the glob change is the right extra (a glob is not a path either). I did not Approve — leaving the duplicate/sibling pick to maintainers.

alt-glitch:

Related: #80407 covers the same Windows regex/glob quoting fault, but is dirty and predates native-rg path handling. This mergeable branch retains the newer path conversion and adds the matching glob coverage, so it is a current landing candidate rather than a duplicate.

PR #98623, opened 2026-08-30
fix(tools): skip capability-incapable backends in web auto-detect

### LINKED ISSUE #98618: [Bug]: _is_backend_available() ignores capability — search-only backends report available for extract
### Bug Description

`_is_backend_available()` in `tools/web_tools.py` answers "is this backend reachable?" but is used to answer "can this backend do the thing I'm about to ask?" It takes no capability argument, so a search-only backend is reported available for **extract**, and the failure surfaces only at call time.

`_get_capability_backend(capability)` receives the capability and then drops it:

```python
def _get_capability_backend(capability: str) -> str:
    cfg = _load_web_config()
    specific = (cfg.get(f"{capability}_backend") or "").lower().strip()
    if specific and _is_backend_available(specific):   # capability not passed
        return specific
    return _get_backend()
```

`_get_backend()`'s auto-detect cascade has the same gap — it selects `ddgs` on package importability alone, with no capability check, and is shared by both search and extract.

The signal already exists and is correct: `WebSearchProvider` declares `supports_search()` / `supports_extract()`, and `plugins/web/ddgs/provider.py` returns `False` for extract. The resolver simply never consults it.

### Steps to Reproduce

With no `TAVILY_API_KEY` / `EXA_API_KEY` / `PARALLEL_API_KEY` / `FIRECRAWL_API_KEY` set, `web.*_backend` empty, and `ddgs` installed:

```python
from tools.web_tools import _get_extract_backend, _is_backend_available, web_extract_tool
from agent.web_search_registry import get_provider
import asyncio

b = _get_extract_backend()
print(b)                                    # ddgs
print(_is_backend_available(b))             # True   <-- claims usable
print(get_provider('ddgs').supports_extract())  # False  <-- provider disagrees
print(asyncio.run(web_extract_tool(['https://example.com'])))
# {"success": false, "error": "DuckDuckGo (ddgs) is a search-only backend
#  and cannot extract URL content. ..."}
```

### Expected

`_get_extract_backend()` should skip backends whose `supports_extract()` is `False` and fall through to one that can extract, or report no extract backend available. A backend reported available for a capability should be able to serve it.

### Actual

Extract resolves to a search-only backend, availability reports `True`, and the call fails at runtime.

### Impact

The misreport is not limited to `ddgs`. In the same environment I saw `_get_backend()` return `firecrawl` while `_is_tool_gateway_ready()` was `False` — availability and usability disagreeing again, by a different route. Because the docstring calls `_is_backend_available` "the single chokepoint through which `_get_backend`, `_get_capability_backend`, and `check_web_api_key` all resolve availability," any caller trusting it inherits the gap — including setup/doctor surfaces that tell users a backend is configured and working when extract cannot run.

### Suggested Fix

Thread the capability through, e.g. `_is_backend_available(backend, capability=None)`, and when `capability` is given consult the registered provider's `supports_<capability>()`. Built-in backends can carry a small static capability map. `_get_backend()` would take the same optional parameter so the auto-detect cascade skips capability-incapable candidates.

### Related but distinct

#88116 and #73192 cover plugin **load-ordering** on cold start (registry empty at selection time). This is orthogonal: it reproduces with everything loaded, because the capability is never consulted at all.

### Environment

Fork of `NousResearch/hermes-agent` (thin fork; `tools/web_tools.py` unmodified from upstream), Python 3.11, Linux/WSL2.

liuhao1024:
## What does this PR do?

`_get_backend()` resolved both search and extract through the same availability ladder, so on an install with the `ddgs` package importable and no extract-capable credentials, `web_extract` resolved to `ddgs` — a search-only backend — and only failed at call time with the "search-only backend" error. Availability was answering "is this backend reachable?" where the caller meant "can this backend do the thing I'm about to ask?"

This threads the capability through backend selection:

- New `_backend_supports_capability(backend, capability)` helper: registered providers answer via their own `supports_search()`/`supports_extract()`; names with no registered provider (plugins not yet loaded — subprocess agent runs, delegate children, scripts) fall back to a small static search-only map (`searxng`, `brave-free`, `ddgs`, `xai`). Unknown backends are assumed capable so a custom provider's strict selection is never silently dropped.
- `_get_backend(capability=None)`: the credential ladder, plugin-registered walk, and keyless walk now skip backends that cannot serve the requested capability. Explicit config (`web.backend` / `web.{capability}_backend`) stays strictly returned as-is — the dispatch path's typed search-only error still surfaces misconfiguration honestly.
- `_is_backend_available(backend, capability=None)` gains the same optional gate so callers asking the chokepoint about a capability get an honest answer.

After the fix, extract on a ddgs-only install falls through to the `firecrawl` default (surfacing Firecrawl's own setup error when unconfigured) instead of pretending ddgs will work; search still resolves to ddgs.

## Related Issue

Fixes #98618

## Type of Change

## Changes Made

- `tools/web_tools.py`: added `_SEARCH_ONLY_WEB_BACKENDS` and `_backend_supports_capability()`; `_get_backend()` takes an optional `capability` and filters the credential ladder / plugin walk / keyless walk; `_get_capability_backend()` passes its capability through; `_is_backend_available()` takes an optional `capability` gate.
- `tests/tools/test_web_tools_config.py`: new `TestCapabilityAwareSelection` covering ddgs/searxng being skipped for extract while search keeps them, extract-capable backends not being dropped, registry-priority vs static-map capability resolution, and the `_is_backend_available` capability gate.

## How to Test

1. `python -m pytest tests/tools/test_web_tools_config.py::TestCapabilityAwareSelection -q` — all 6 tests should pass.
2. Reproduce the original report (no `TAVILY_API_KEY`/`EXA_API_KEY`/`PARALLEL_API_KEY`/`FIRECRAWL_API_KEY` set, no `web.*_backend` config, `ddgs` importable):
   ```python
   from tools.web_tools import _get_extract_backend, _get_search_backend
   _get_extract_backend()  # "firecrawl" (was "ddgs")
   _get_search_backend()   # "ddgs" (unchanged)
   ```
   Observed result: before the fix `_get_extract_backend()` returned `ddgs` and `web_extract_tool` failed with the search-only error; after the fix extract resolves to `firecrawl` and search is unaffected.
3. `python -m pytest tests/tools/test_web_tools_config.py tests/tools/test_web_providers.py tests/tools/test_web_providers_ddgs.py tests/tools/test_web_providers_xai.py tests/tools/test_web_keyless_fallback.py tests/tools/test_web_keyless_rescue.py -q` — all pass (2 pre-existing local env failures in `TestParallelClientConfig` reproduce on clean upstream/main and are unrelated).

### Code

### DIFF
--- a/tests/tools/test_web_tools_config.py
+++ b/tests/tools/test_web_tools_config.py
@@ -352,6 +352,80 @@ class TestBackendSelection:
             assert _get_backend() == "firecrawl"
 
 
+class TestCapabilityAwareSelection:
+    """Auto-detect must skip backends that cannot serve the capability (#98618).
+
+    With ddgs installed and no extract-capable credentials, extract used to
+    resolve to the search-only ddgs: selection reported it "available" and
+    the failure only surfaced at call time with the search-only error.
+    """
+
+    def test_extract_skips_search_only_ddgs(self):
+        """ddgs importable + no credentials → extract falls through to the
+        firecrawl default instead of search-only ddgs; search keeps ddgs."""
+        from tools.web_tools import _get_extract_backend, _get_search_backend
+        with patch("tools.web_tools._load_web_config", return_value={}), \
+             patch("tools.web_tools._is_tool_gateway_ready", return_value=False), \
+             patch("tools.web_tools._ddgs_package_importable", return_value=True), \
+             patch("tools.web_tools._list_registered_web_providers", return_value=[]), \
+             patch("agent.web_search_registry._keyless_tier_enabled", return_value=False):
+            assert _get_extract_backend() == "firecrawl"
+            assert _get_search_backend() == "ddgs"
+
+    def test_extract_prefers_extract_capable_tavily_over_ddgs(self):
+        """Capability filtering must not drop extract-capable backends:
+        a Tavily key still wins over an importable ddgs for extract."""
+        from tools.web_tools import _get_extract_backend
+        with patch("tools.web_tools._load_web_config", return_value={}), \
+             patch("tools.web_tools._ddgs_package_importable", return_value=True), \
+             patch.dict(os.environ, {"TAVILY_API_KEY": "tvly-test"}):
+            assert _get_extract_backend() == "tavily"
+
+    def test_extract_skips_searxng(self):
+        """A search-only credential (SEARXNG_URL) never satisfies extract —
+        it falls through to the firecrawl default."""
+        from tools.web_tools import _get_extract_backend
+        with patch("tools.web_tools._load_web_config", return_value={}), \
+             patch("tools.web_tools._is_tool_gateway_ready", return_value=False), \
+             patch("tools.web_tools._ddgs_package_importable", return_value=False), \
+             patch("tools.web_tools._list_registered_web_providers", return_value=[]), \
+             patch("agent.web_search_registry._keyless_tier_enabled", return_value=False), \
+             patch.dict(os.environ, {"SEARXNG_URL": "http://searxng.local"}):
+            assert _get_extract_backend() == "firecrawl"
+
+    def test_capability_check_uses_registered_provider(self):
+        """A registered provider's supports_extract() wins over any static map."""
+        from tools.web_tools import _backend_supports_capability
+        provider = MagicMock()
+        provider.supports_search.return_value = True
+        provider.supports_extract.return_value = False
+        with patch("tools.web_tools._registered_web_provider", return_value=provider):
+            assert _backend_supports_capability("custom-backend", "search") is True
+            assert _backend_supports_capability("custom-backend", "extract") is False
+
+    def test_capability_check_static_map_without_registry(self):
+        """Registry absent (plugins not loaded) → the static search-only map
+        applies; unknown backends stay extract-capable so a custom provider's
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Fixes auto-detect selecting search-only backends for `extract` capability (#98618). `ddgs` importable previously resolved `extract` to `ddgs` and failed at call time.

**Correctness:**
- `tests/tools/test_web_tools_config.py:12-80` — `TestCapabilityAwareSelection` covers: `extract` skips `ddgs` → `firecrawl`, `tavily` (extract-capable) still wins over `ddgs`, `searxng` search-only skipped, `_backend_supports_capability` with registered provider vs static map, `_is_backend_available(capability="extract")` gate. Static map correctly marks `ddgs`/`brave-free`/`searxng` search-only, unknown → extract-capable (fail-open for custom providers).
- Implementation (not shown in diff excerpt but exercised by tests) adds `_backend_supports_capability` and `_get_extract_backend` filtering — tests are the spec.

**Non-blocking:** None.

**Verdict:** LGTM.

liuhao1024:
Thanks for the review — glad the capability-filtering tests read as the spec here; the fail-open default for unknown custom providers is the intended behavior.

PR #54610, opened 2026-06-29
fix(cli): detect basic plugin disabled when dashboard auth configured (#54489)

AlexFucuson9:
## Problem

When `basic` is in `plugins.disabled` but `dashboard.basic_auth` is configured, `hermes dashboard --host 0.0.0.0` fails with a generic "no auth providers registered" error. The root cause is that `hermes setup` (blank-slate mode) can add `basic` to `plugins.disabled`, silently breaking the password auth provider.

Users see:
```
Refusing to bind dashboard to 0.0.0.0 — the auth gate engages on non-loopback binds, but no auth providers are registered.
```

No hint that `plugins.disabled` is the cause.

## Fix

Two targeted changes in 2 files:

1. **`hermes_cli/web_server.py`**: When no auth providers are found on a non-loopback bind, check if `basic` is in `plugins.disabled` while `dashboard.basic_auth.username` is configured. If so, surface a specific diagnostic with actionable remediation:
   ```
   • basic: plugin is disabled in plugins.disabled but dashboard.basic_auth is configured.
     Fix: remove 'basic' from plugins.disabled in config.yaml (or run: hermes plugins enable basic)
   ```

2. **`hermes_cli/doctor.py`**: Add a check in the "Auth Providers" section that warns about this misconfiguration so users can detect it proactively via `hermes doctor`.

Fixes #54489

Issue #54489, opened 2026-06-28
hermes setup disables basic plugin → dashboard auth fails silently on non-loopback bind (v0.17.0)


JHuette:
### Bug Description

hermes setup adds the `basic` plugin to plugins.disabled. Afterwards, hermes dashboard --host 0.0.0.0 fails with "no auth providers registered" even though dashboard.basic_auth is correctly configured in config.yaml.

### Steps to Reproduce

1. Run: hermes setup (full wizard)
2. Configure dashboard.basic_auth in config.yaml (username + password_hash)
3. Run: hermes dashboard --host 0.0.0.0

### Expected Behavior

Dashboard starts with basic auth gate active.

### Actual Behavior

Refusing to bind dashboard to 0.0.0.0 — no auth providers are registered.
No hint that plugins.disabled is the cause.

### Affected Component

Configuration (config.yaml, .env, hermes setup)

### Debug Report

```shell
https://paste.rs/UqC6R
```

### Operating System

Ubuntu 26.04 LTS

### Hermes Version

0.17.0

### Additional Logs / Traceback (optional)

```shell
hermes doctor does not warn when dashboard.basic_auth is configured but the basic plugin is disabled.
Remove `basic` from plugins.disabled and add it to plugins.enabled in config.yaml.
```

### Are you willing to submit a PR for this?

- [ ] I'd like to fix this myself and submit a PR

hoekbrwr:
### Same symptom, additional root cause

We encountered this crash on a BigBear CasaOS Hermes container (v0.17.0) on ZimaOS. In our case the **basic plugin was NOT disabled** — it was correctly configured in `dashboard.basic_auth` with username + password_hash, and the plugin was in `plugins.enabled`.

The crash-loop was caused by a secondary bug: even with the basic plugin enabled, the middleware auth gate check `if len(providers) != 1` in `dashboard_auth/middleware.py` rejects binding to `0.0.0.0` because `BasicAuthProvider.start_login()` raises `NotImplementedError`, so the providers list is effectively empty at the gate-check point.

### Our working fix

Applied this patch to `dashboard_auth/middleware.py`:

```python
# Before:
if len(providers) != 1:

# After:
if len(providers) != 1 or any(getattr(p, "supports_password", False) for p in providers):
```

This allows binding to `0.0.0.0` when a password-based auth provider (like BasicAuthProvider) is present, even if its `start_login()` is a stub/NotImplemented.

### Repro steps

[...]

PR #98624, opened 2026-08-30
fix(dashboard): run plugin discovery before the auth gate

z10g:
## Problem

With `dashboard.basic_auth` correctly configured in `config.yaml` (and the bundled `basic` plugin enabled), `hermes dashboard --host 0.0.0.0` still refuses to start on v0.20.6 and current `main`:

```
Refusing to bind dashboard to 0.0.0.0 — the auth gate engages on non-loopback binds, but no auth providers are registered.
```

## Root cause

The gate in `start_server()` (`hermes_cli/web_server.py`) checks `list_providers()` and fails closed when it's empty. But auth providers register through the plugin hook (`ctx.register_dashboard_auth_provider`), which only runs during plugin discovery — and the dashboard startup path never triggers discovery before this check.

The only `discover_plugins()` calls in `web_server.py` live in request-time helpers (channels-page rendering, terminal-backend picker), so at gate-check time the provider registry is empty unless some unrelated import happened to trigger discovery as a side effect. Notably, the gate's own diagnostics import `plugins.dashboard_auth.nous` to read its `LAST_SKIP_REASON` — the plugin module is importable right there, but its `register()` hook never ran.

This is the "same symptom, additional root cause" case reported by @hoekbrwr in #54489 (plugin enabled + credentials configured, still refused), and is distinct from the `plugins.disabled` diagnostics proposed in #54610.

## Fix

Trigger `discover_plugins()` right before the provider check. It is idempotent (it joins any background discovery first), matching the existing idiom already used elsewhere in this file. The gate keeps failing closed when discovery registers no providers.

## Testing

- Added two regression tests in `tests/hermes_cli/test_dashboard_auth_gate.py`:
  - the gate runs discovery and proceeds when a provider registers during it;
  - the gate still fails closed when discovery runs but registers nothing.
- Verified end to end on device (Termux/aarch64, v0.20.6 + this patch): dashboard binds on `0.0.0.0:9119`, the login page serves, and password login via `POST /auth/password-login` succeeds with configured credentials.

### DIFF
--- a/hermes_cli/web_server.py
+++ b/hermes_cli/web_server.py
@@ -19526,6 +19526,17 @@ def start_server(
         # provider to be registered, else fail closed — there is no longer an
         # escape hatch that serves the dashboard without authentication.
         from hermes_cli.dashboard_auth import list_providers
+        # Plugin-registered auth providers (e.g. the bundled basic password
+        # provider) only register during plugin discovery, which this server
+        # process doesn't otherwise run before this gate — trigger it
+        # explicitly (idempotent) or a fully-configured provider stays
+        # invisible here and the gate fails closed.
+        try:
+            from hermes_cli.plugins import discover_plugins
+
+            discover_plugins()
+        except Exception:
+            _log.warning("plugin discovery before auth gate failed", exc_info=True)
         if not list_providers():
             # Surface the *specific* reason any bundled provider declined
             # to register (e.g. missing HERMES_DASHBOARD_OAUTH_CLIENT_ID).
--- a/tests/hermes_cli/test_dashboard_auth_gate.py
+++ b/tests/hermes_cli/test_dashboard_auth_gate.py
@@ -479,3 +479,78 @@ def test_should_require_dashboard_auth_truth_table(
     else:
         monkeypatch.setenv("HERMES_DASHBOARD_PUBLIC_URL", public_url)
     assert should_require_dashboard_auth(host) is expected
+
+
+# ---------------------------------------------------------------------------
+# Plugin discovery before the auth gate
+# ---------------------------------------------------------------------------
+
+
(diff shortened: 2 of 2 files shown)


Enough1122:

**Overall:** Fixes dashboard auth gate failing closed despite valid `dashboard.basic_auth` because plugin-registered providers weren't discovered before the gate.

**Correctness:**
- `hermes_cli/web_server.py:10-15` — calls `discover_plugins()` (idempotent) before `list_providers()`, wrapped in `try/except` with `warning` + `exc_info`. Ensures bundled `basic` password provider registers in server process that otherwise never runs discovery before gate.
- Test `test_start_server_auth_gate_runs_plugin_discovery` fakes `discover_plugins` to register `StubAuthProvider`, asserts `calls == [1]` and `auth_required == True` plus `uvicorn` host — proves discovery ran and gate satisfied.

**Non-blocking nit:**
- `discover_plugins()` may have side effects beyond auth providers; calling it earlier is safe due to idempotence claim, but verify it doesn't double-register other providers (should be guarded).

**Verdict:** LGTM.

z10g:
Thanks for the review! Re the non-blocking nit — double-registration is already guarded:

- `discover_and_load()` short-circuits under `_discovery_lock` when `self._discovered` is set (`plugins.py:4234`), so each plugin's `register()` runs at most once per process unless `force=True`. The flag is even set up front as a re-entrancy guard, since a plugin's own `register()` can transitively trigger discovery again (`plugins.py:4245-4250`); `force=True` goes through `unload()` so the ledger tears down prior state first.
- This matches existing upstream usage — `web_server.py` already calls `discover_plugins()` per request in the channels-page and terminal-backend helpers, which would have double-registered long ago if it weren't safe.

PR #98593, opened 2026-08-30
feat(zai): add Responses API provider support

ecourn:
## Summary

Z.AI serves GLM models on a native OpenAI **Responses** wire at `https://api.z.ai/api/v1` (`POST /v1/responses`), separate from the Chat Completions endpoint (`/api/paas/v4`) already covered by the built-in `zai` profile. Until now, reaching the Responses wire required a hand-written `providers:` entry in `config.yaml` with `transport: codex_responses`. This PR registers it as a first-class plugin profile so `--provider zai-responses` works out of the box.

The change is **purely additive** (3 new files, no existing file touched):

- `plugins/model-providers/zai-responses/` — declarative `ProviderProfile` (`api_key` auth, `api_mode="codex_responses"`), same pattern as the `xai` profile. Auto-wires into `PROVIDER_REGISTRY`, aliases, `CANONICAL_PROVIDERS`, doctor and the picker via the existing registry auto-extension (no edits to `auth.py` / `runtime_provider.py` needed).
  - Keys: `GLM_API_KEY` (Z.ai's documented variable, shared with the `zai` profile) with `ZAI_API_KEY` / `Z_AI_API_KEY` aliases; `ZAI_RESPONSES_BASE_URL` overrides the endpoint for relays.
  - `fetch_models` override parses Z.ai's `{"models": [...]}` / `slug` catalog shape — the generic `{"data": [...]}` / `id` parser returns nothing on a healthy Z.ai response.
  - `fallback_models` pinned to the live catalog: `glm-5.3`, `glm-5.3-flash`, `glm-5-turbo`.
- `tests/hermes_cli/test_zai_responses_provider.py` — 21 behavior contracts (registration, alias resolution, registry auto-wiring, wire-mode detection, catalog parsing incl. `supported_in_api: false` filtering, bearer-auth fetch via stubbed opener). All offline, no network, no real keys.

The existing `zai` profile, its aliases, and the `api.z.ai` reverse hostname mapping are untouched; user `providers:` config still takes precedence, so nothing changes for existing setups.

## Live validation

Verified against the real endpoint (2026-08-30):

- `POST https://api.z.ai/api/v1/responses` with `model: glm-5.3-flash` → HTTP 200 (streaming tool-calling turn).
- `GET https://api.z.ai/api/v1/models` → HTTP 200, `{"models": [{"slug": ..., "supported_in_api": ...}]}` envelope (the shape the parser handles).
[...]

PR #98627, opened 2026-08-30
fix(cli): prevent setproctitle decode failures from aborting startup

### LINKED ISSUE #98620: _set_process_title crashes with UnicodeDecodeError when argv ends in multibyte chars (setproctitle truncation)
### Environment

- macOS (arm64), Python 3.11.15, setproctitle 1.3.7 (latest on PyPI)
- hermes-agent v0.20.6

### Symptom

`hermes -p <profile> chat -q '<any prompt ending in a multibyte char (Chinese/emoji)>'` crashes at startup:

```
File "hermes_cli/main.py", line 13108, in main -> _set_process_title()
File "hermes_cli/main.py", line 251 -> import setproctitle
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 310-311: unexpected end of data
```

ASCII-only argv and `hermes --version` are unaffected. The crash kills the process at the very first statement of main(), so the agent never starts.

### Root cause

setproctitle 1.3.7 on Darwin (PS_USE_CLOBBER_ARGV): when environ is not contiguous with the argv region, `ps_buffer_size` equals the argv region size exactly. `init_ps_display` writes the space-joined title with `spt_strlcpy`, which reserves one byte for the NUL terminator — the last byte of the title is silently cut. If the last argument ends with a multibyte character, the cut lands mid-sequence, and the module-level `getproctitle()` call at import (setproctitle/__init__.py:56-57, the #113 fork workaround) decodes the truncated buffer via `Py_BuildValue("s#")` -> UnicodeDecodeError.

Deterministic repro: `SPT_NOENV=1` forces the non-contiguous-environ branch (setproctitle skips extending the buffer over environ), so the crash always occurs.

### Suggested fix

In `_set_process_title` (main.py:249-256), widen the guard so a cosmetic title-setter can never kill the agent:

```python
except (ImportError, UnicodeDecodeError):
    pass
```

The ctypes fallback (strategy 2/3) already handles the title gracefully, so the only behavior change is that a crash in the optional dependency no longer aborts startup.

### Repro

```
SPT_NOENV=1 /path/to/venv/bin/hermes model <80 Chinese characters>
# -> UnicodeDecodeError at import setproctitle
```

Related upstream issue: dvarrazzo/py-setproctitle (Darwin strlcpy multibyte-tail truncation), filed in parallel.

ryrenz:
## Summary

Hermes could exit before entering the CLI main loop on macOS when `setproctitle` decoded an argv buffer truncated through a multibyte character. `_set_process_title()` only treated a missing optional dependency as non-fatal, allowing an import-time `UnicodeDecodeError` to escape.

## Changes

- Treat `UnicodeDecodeError` during the optional `setproctitle` strategy as a best-effort title failure and continue to the existing ctypes fallback.
- Add a hermetic regression test that reproduces the import-time decoding failure without invoking a real platform title syscall.

The change is limited to process-title initialization and does not alter normal title-setting behavior.

## Validation

- `scripts/run_tests.sh tests/cli/test_set_process_title.py -q`
- `git diff --check`

Fixes #98620

### DIFF
--- a/hermes_cli/main.py
+++ b/hermes_cli/main.py
@@ -193,6 +193,16 @@ def _set_process_title() -> None:
         return
     except ImportError:
         pass
+    except UnicodeDecodeError:
+        # A corrupt setproctitle .pyc raises this out of the import machinery, and an
+        # unhandled one here aborts startup entirely (#98593). Degrade to the ctypes
+        # fallback, but say so once — a silent permanent no-op is how this was found.
+        import logging as _logging
+
+        _logging.getLogger(__name__).debug(
+            "setproctitle import failed to decode (corrupt bytecode?); using the ctypes fallback",
+            exc_info=True,
+        )
 
     import ctypes
     import platform
--- a/tests/cli/test_set_process_title.py
+++ b/tests/cli/test_set_process_title.py
@@ -1,0 +1,44 @@
+"""Regression tests for the best-effort process title helper."""
+
+import builtins
+from unittest.mock import patch
+
+from hermes_cli.main import _set_process_title
+
+
+def test_set_process_title_survives_unicode_decode_error_during_import():
+    real_import = builtins.__import__
+
+    def import_with_decode_failure(name, *args, **kwargs):
+        if name == "setproctitle":
+            raise UnicodeDecodeError(
+                "utf-8", b"\xe4", 0, 1, "unexpected end of data"
+            )
+        return real_import(name, *args, **kwargs)
+
+    with (
+        patch("builtins.__import__", side_effect=import_with_decode_failure),
+        patch("platform.system", return_value="Windows"),
+    ):
+        _set_process_title()
+
+
+def test_set_process_title_logs_once_when_bytecode_is_corrupt(caplog):
+    """The fallback must not be silent — a permanent no-op is how #98593 was found."""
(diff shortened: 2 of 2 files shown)


kokhlo:
Reading this alongside the call sites — catching `UnicodeDecodeError` at the `import setproctitle` boundary is the right first layer for #98593. One defensive gap worth considering before merge:

the adjacent call site (`hermes_cli/main.py:13108`, `_set_process_title()`) is a silent best-effort with no logging anywhere on the fallback path. If the `.pyc` corruption is environmental (failing disk, stray `PYTHONDONTWRITEBYTECODE` interaction), the feature degrades to a permanent no-op and nothing in the logs or `hermes doctor` ever says so — the operator is back to discovering it the way this issue was discovered.

A one-time DEBUG/WARN line when the fallback path is taken (not per-call — it fires on every process title update) plus a small doctor check would make the degradation visible. Not blocking; the catch itself is correct and well-scoped.

ryrenz:
Good call — added. The fallback now emits a one-shot DEBUG line (with `exc_info`) naming the decode failure, so a permanently degraded title is visible in logs instead of silent. Covered by a second test asserting the record is emitted. Also rebased onto current main.

PR #98629, opened 2026-08-30
fix(desktop): stop session.info heartbeats re-rendering every tab, and stop replaying approvals against a reaped runtime

Dolverin:
Two related desktop fixes for the same symptom: a multi-tab session that
flickers and eats keystrokes. Both were found by reading this install's own
logs, and the numbers below come from those logs, not from a synthetic repro.

## 1. `session.info` heartbeats published a new state object every time

The backend stamps `model` / `provider` / `cwd` / `fast` / … on **every**
`session.info`, and those arrive in bursts. Each one spread a fresh
`ClientSessionState` into `$sessionStates` — the rendering source for every
mounted session surface — so every burst re-rendered all of them. The cost
scaled with the number of open tabs, which is exactly how the flicker
presented.

`applySessionInfoStatePatch` now returns the **same** state object when all
nine patch fields already match, so an unchanged heartbeat publishes nothing.

This sits on top of `0401e0888`, which fixed the neighbouring stranger's-cwd
flicker in the same file.

## 2. Approval replay never stopped against a reaped runtime

`replayPendingApproval` fires on every `session.info`, and the caller discards
failures with `.catch(() => undefined)`. When the gateway no longer holds a
runtime it answers `4001 "session not found"` and expects the client to resume
the stored session instead — but nothing read that refusal, so a single
detached tab polled `approval.pending` forever.

`logs/gui.log` on this install recorded **5665 refusals at a steady 33-36/min
for ~22h**, each one a gateway-side WARNING, all for the same dead runtime id.

The refusal is now read before it is discarded, and that runtime id stops
being polled. The mark is about the *runtime*, not the conversation: a resume
comes back under a new id, which is not in the set, and `gateway.ready` clears
the set wholesale since a reconnect rebuilds the gateway's runtime map.

`isDetachedSessionRpc` is deliberately narrow (`4001` / "session not found") so
a socket drop or timeout cannot latch a live session out of its recovery — a
regression test pins that.

## Tests

```
vitest run --project ui \
  src/app/session/hooks/use-message-stream/gateway-event/session-info.test.ts \
  src/store/prompts.test.ts

Test Files  2 passed (2)
     Tests  22 passed (22)
```

`npm run typecheck` clean (all three tsconfigs).

The identity regression test is red without the first fix and green with it.

### DIFF
--- a/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/index.ts
+++ b/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/index.ts
@@ -10,7 +10,7 @@ import {
 import { reconcileSessionCompacting } from '@/store/compaction'
 import { $gateway, activeGatewayConnectionId } from '@/store/gateway'
 import { $activeGatewayProfile, normalizeProfileKey } from '@/store/profile'
-import { replayPendingApproval } from '@/store/prompts'
+import { forgetDetachedApprovalSessions, replayPendingApproval } from '@/store/prompts'
 import { setSessionProviderWait } from '@/store/provider-wait'
 import { setSessionDraftingTool } from '@/store/tool-drafting'
 import type { RpcEvent } from '@/types/hermes'
@@ -194,6 +194,12 @@ export function useGatewayEventHandler(deps: GatewayEventDeps) {
       }
 
       const isActiveEvent = !!sessionId && sessionId === activeSessionIdRef.current
+
+      // A reconnect rebuilds the gateway's runtime map, so every runtime we
+      // had written off as detached deserves one more look.
+      if (event.type === 'gateway.ready') {
+        forgetDetachedApprovalSessions()
+      }
 
       const replaySessionId = approvalReplaySessionId(event.type, activeSessionIdRef.current, sessionId)
 
--- a/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/session-info.test.ts
+++ b/apps/desktop/src/app/session/hooks/use-message-stream/gateway-event/session-info.test.ts
@@ -1,5 +1,7 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
+import type { ClientSessionState } from '@/app/types'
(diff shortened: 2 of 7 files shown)


Dolverin:
CI is red on one test, and I believe it is a flake rather than this PR. I
cannot re-run the job myself (no admin rights on this repo), so here is the
evidence for whoever can:

**The failure**

```
FAIL  ui  src/app/settings/gateway-settings.test.tsx > GatewaySettings
          > loads the machine-level connection config (no profile scoping)
Error: Test timed out in 15000ms.
 ❯ src/app/settings/gateway-settings.test.tsx:39:3

Test Files  1 failed | 682 passed (683)
     Tests  1 failed | 6756 passed (6757)
```

**Why it is unrelated to this branch**

1. It is a *timeout*, not an assertion. The test body is
   `await import('./gateway-settings')`, and that run reported
   `import 1298.04s` / `environment 674.63s` against a 15s per-test limit —
   the runner was heavily loaded.
2. There is no import path from anything this PR touches.
   `gateway-settings.tsx` imports none of the changed modules
   (`gateway-event/index.ts`, `session-info.ts`, `use-message-stream/utils.ts`,
   `lib/gateway-rpc.ts`, `store/prompts.ts`), and the test file mocks its own
   dependencies.
3. It passes on this branch locally:

```
npx vitest run --project ui src/app/settings/gateway-settings.test.tsx
Test Files  1 passed (1)
     Tests  1 passed (1)
  Duration  10.24s
```

4. `check:lint` (typecheck + eslint) passed in the same CI run, and passes
   locally: 0 errors, and none of the 124 pre-existing warnings are in files
   this PR touches.

A re-run should clear it. If the test is flaky more broadly, raising its
timeout or pre-importing the module would be a separate change — happy to
open one if that is wanted.

OutThisLife:
Superseded by #99891.

That salvage keeps the `session.info` same-object publish and the approval-replay backoff, and uses the shared `runtime-gone` latch instead of a private `detachedApprovalSessions` set.

You're credited via `Co-authored-by`. Thanks for the heartbeat identity fix.

PR #96086, opened 2026-08-27
fix(gateway): keep raw-session delegation completions recoverable

fangliquanflq:
## What does this PR do?

Async delegations completed for raw sessions could become permanently unrecoverable when a gateway started without an `api_server` adapter. This change keeps those durable completion rows pending without consuming delivery attempts, so a later gateway start can restore and retry the wake instead of exhausting the retry budget across restarts.

### Symptom

The gateway logged `Dropping watch notification for raw session ...: no api_server adapter to self-post through` after the delegated work had completed. No notification reached the originating raw session.

### Impact

Users could lose visibility of completed background delegation work. Repeated gateway starts consumed the bounded delivery-attempt budget even though no delivery transport existed, eventually marking the durable completion as dropped.

### Bug Cause

**Trigger:** `gateway/run.py:_deliver_completion_notification` when an async delegation event has an unstructured raw session id and `Platform.API_SERVER` is absent.

**Causal chain:**
1. Startup restoration re-enqueues the durable pending completion.
2. `_deliver_completion_notification` claims the row before checking whether this gateway has a transport for the raw session, incrementing `delivery_attempts`.
3. `_inject_watch_notification` finds no `api_server` adapter, returns without delivery, and releases the claim. Each restart repeats this until the retry cap marks the row dropped.

**Why it is wrong:** The retry budget represents failed delivery attempts, but no adapter means no delivery attempt was made.

**Working sibling / contrast:** A configured `api_server` adapter self-posts the wake and acknowledges the durable row after acceptance. Transient adapter failures remain retryable and continue consuming the bounded attempt budget.

**Ruled out:** The completion was not missing from persistence. The durable row and startup restore path already existed; the loss came from claiming it before transport availability was known.

### Fix

Detect durable raw-session events before claiming their SQLite delivery row. If this gateway has no `api_server` adapter, leave the row untouched and pending for startup restoration in a future lifecycle. The raw-session id extraction is shared with the existing self-post path to keep both decisions consistent.

## Related Issue

Closes #96070

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `gateway/run.py` - defer durable raw-session completion claims until an `api_server` transport exists.
- `tests/gateway/test_completion_delivery.py` - prove the event remains pending with zero delivery attempts and is restored on a later startup.

## How to Test

1. Persist a completed async delegation whose session key is a raw session id.
2. Drain it through a gateway runner without an `api_server` adapter.
[...]

PR #82704, opened 2026-08-09
fix(gateway): defer unroutable async completions

eventh0riz0n:
## What does this PR do?

Prevents a messaging gateway from consuming durable async-delegation delivery attempts when that gateway cannot prove it owns the completion's route.

The read-only preflight now mirrors the real injector:

- persisted or cached `SessionSource` is authoritative even when the legacy key is opaque;
- logical Slack/Discord routes fronted by `Platform.RELAY` are recognized through the same alias-aware resolver used by delivery;
- opaque sessions require an explicit `origin_session_id` stamp before stateless api_server delivery is authorized;
- a coalesced batch proves route ownership and claims its primary row before any siblings, so an unowned batch leaves every row `pending` with `delivery_attempts == 0`;
- unexpected resolver failures fall back to the established claim/deliver/release path rather than hot-looping before the durable retry budget can advance.

Returning `None` leaves SQLite as the durable inbox while avoiding an in-memory retry loop during that gateway lifetime.

### 2026-09-03 refresh

Rebased onto current `main`. The API-server completion regression now follows the current client-owned-turn contract: an explicitly owned async delegation is persisted as a durable delivery row rather than self-posted as a new user turn. Focused and adjacent suites pass 97/97.

## Related Issue

Fixes #82703

Related: #62479, #75117, #71908, #79829. This is the narrow current-main ledger invariant; it does not replace broader origin/profile work in those PRs.

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `gateway/run.py`
  - recover persisted/cached structured origins before legacy-key parsing;
  - use `resolve_delivery_transport()` for alias-aware ownership preflight;
  - establish ownership and primary durable claim before claiming coalesced siblings;
  - pass an already-established primary claim through the normal delivery/ack/release seam.
- `tests/gateway/test_completion_delivery.py`
  - cover opaque persisted routes, relay-only Slack/Discord routes, stamped/unstamped api_server ownership, unowned multi-event batches, retries, competing claims, and adapter acceptance/ack ordering.

## How to Test

```bash
python -m pytest -q -o addopts= \
  tests/gateway/test_completion_delivery.py \
[...]

PR #81806, opened 2026-08-08
feat(api_server): expose session async delegations over HTTP

meiqinsi:
External /v1/runs orchestrators treat run.completed as job-done, but background delegate_task children can outlive the parent turn. Add GET /api/sessions/{id}/delegations as a thin projection of the async registry so clients can poll until session work actually drains. Closes #81754

## What does this PR do?
External runtimes that drive Hermes via `POST /v1/runs` + SSE often treat
`run.completed` as "the job is finished". That is correct for the **parent
turn**, but `delegate_task(background=true)` children can keep running after
the parent run terminates; their results re-enter the session later as a wake
turn.
This PR adds `GET /api/sessions/{session_id}/delegations` so orchestrators
can poll until session async work is **fully drained** — without redefining
`run.completed` semantics.
### Two drain phases (why both signals matter)
| Phase | Signal | Meaning |
|-------|--------|---------|
| Subagent live work | `active_count` | In-memory registry: `running` / `stalling` / `finalizing` |
| Wake delivery | `pending_delivery_count` | Terminal delegations with `delivery_state='pending'` in durable SQLite |
| Fully drained | `drain_complete` | `active_count == 0 && pending_delivery_count == 0` |
**Production note:** A subagent can finish (`active_count → 0`) while the
parent session is still processing the wake turn (`pending_delivery_count > 0`).
Polling only `active_count == 0` is insufficient for external orchestrators
that finalize after background work; they should poll until `drain_complete`.
This is a **read-only API projection** of existing durable `delivery_state`
rows. It does **not** change wake injection, `gateway/run.py`, or agent turn
semantics.

## Related Issue

Fixes #81754

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [x] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

### Core (initial)
- `tools/async_delegation.py`: `list_async_delegations_for_session()` — match
  `origin_session_id` / `session_key` / `parent_session_id`; default active-only
- `gateway/platforms/api_server.py`:
  - `GET /api/sessions/{session_id}/delegations`
  - advertise under `/v1/capabilities` (`session_delegations`)
  - whitelist projection + secret redaction on `goal`/`goals` (no `context` / callables)
 
[...]

PR #98630, opened 2026-08-30
fix(gateway): require wake-capable session provenance for background delegate_task

### LINKED ISSUE #98619: api_server: background delegate_task results are undeliverable for header-less OpenAI-compatible clients — wake gate assumes any bound session id is wake-capable
### Bug Description

On `api_server`, `delegate_task(background=true)` from a **header-less OpenAI-compatible client** (plain OpenAI SDK, no `X-Hermes-Session-Id`) dispatches detached subagents whose results are **undeliverable by construction** — the wake gate treats any bound session id as wake-capable, but for fingerprint-derived sessions the wake either hard-fails or lands where the client will never look. The old forced-sync fallback (blocking, but the result arrived in-turn with an honest `note`) is skipped precisely in the cases where it was the only correct behavior.

All refs below are `main@dce2ecb8a`.

### The chain

1. **Header-less requests always get a bound session id.** With no `X-Hermes-Session-Id`, the session id is derived from `sha256(system_prompt + first_user_message)` (`gateway/platforms/api_server.py:5154-5165`, `_derive_chat_session_id:1393`).

2. **The wake gate only checks that a session id exists.** `tools/delegate_tool.py:4188-4209`: when `async_delivery_supported()` is false but `_origin_wake_sid` is non-empty ("the API server always binds one"), it dispatches background and plans delivery via wake self-post — skipping the forced-sync fallback at `:4211-4225`. There is no check that the wake can actually deliver, nor that the client can ever address that session.

3. **Unauthenticated server → wake hard-fails after the work is done.** `gateway/wake.py:116-121` raises `RuntimeError("wake self-post requires API_SERVER_KEY...")` at delivery time. The subagents ran, the parent turn ended long ago, the result is stranded server-side. (`API_SERVER_KEY` is optional for header-less operation, so this is a reachable production config.)

4. **Authenticated server → wake lands in a session the client never reads.** The wake self-POST targets the derived session id, but the header-less request path builds history **from the request body only** — the DB-history branch is header-gated (`api_server.py:5146-5165`). So on the client's next turn the agent has no memory of the wake turn; the delegation results are invisible to the end user. (The derived id is echoed in the `X-Hermes-Session-Id` response header, but plain OpenAI-SDK clients don't read response headers, and sending it back requires `API_SERVER_KEY` anyway — 403 otherwise, `:5115-5129`.)

5. **Sliding-window clients rotate the fingerprint mid-conversation.** Our voice-meeting client trims old transcript from the messages array; when the first user message rotates out, the derived session id silently changes — wakes target the orphaned session, and sandbox/container continuity resets. Any OpenAI-compat frontend that summarizes or trims history hits this.

### How we found it (first-hand ops)

We run hermes-agent on a VPS behind an OpenAI-compatible voice-meeting client (the agent joins a live meeting; a user says "research X while we keep talking"). On v0.19.0 the same call degraded to **forced-sync** (gate then at `tools/delegate_tool.py:3141-3194`): the conversational turn blocked for the entire subagent runtime and only the client's own progress fillers kept the meeting alive. Reading current `main` while preparing this report, we found the successor behavior traded that blocking-but-correct fallback for background-but-lossy in the header-less cases above.

To be explicit about what is measured vs read: the v0.19.0 sync-blocking is a live production observation; the `main` analysis in 1–5 is from code reading at `dce2ecb8a`. Happy to run a repro of any of the three loss cases against our VPS deployment if that helps.

### Suggested direction

At the `delegate_tool.py:4188` gate, treat a session as wake-capable only when the wake can deliver **and** the client can consume it: an explicitly provided `X-Hermes-Session-Id` (not fingerprint-derived — e.g. have `_bind_api_server_session` record the id's provenance) **and** `API_SERVER_KEY` configured. Everything else should keep the forced-sync fallback with its existing honest `note`. Independently, documenting the `X-Hermes-Session-Id`-unlocks-background-delegation contract in the api-server docs would help operators of latency-sensitive clients discover it before production.

### Environment

- hermes-agent v0.19.0 (VPS, Ubuntu 24.04, gateway `api_server` platform) — live observation
- `main@dce2ecb8a` — code analysis
- Client: custom OpenAI-compatible voice-meeting bridge (full-history resend, sliding window)

---

*Found during real VPS operation. Reported by Sho (human) + Alpha (AI agent).*

liuhao1024:
## What does this PR do?

Rebuilt from scratch against the post-refactor `main` (per review #5127266065 and its correction): the chat-completions route now lives in `gateway/platforms/api_server_openai_routes.py`, the wake gate in `tools/delegate_tool_dispatch.py`, and `gateway/session_context.py` was reshaped, so the original patch aimed at code `main` no longer has.

The bug (#98619): a header-less OpenAI-compatible client gets a fingerprint-derived session id bound as the api_server chat id. `delegate_task(background=true)`'s gate treated **any** bound session id as wake-capable and dispatched detached subagents — but for derived ids the wake self-post lands in a session whose history that client never reloads, so the result is undeliverable by construction.

The fix binds a `wake_capable` provenance flag at session-bind time and **default-denies it at the central boundary**: `set_session_vars()` and `_bind_api_server_session()` both treat an omitted declaration as *denied* — a binder that says nothing grants no wake authority. `"1"` is passed only by audited producers whose client can address the bound id again:

- explicit `X-Hermes-Session-Id` on `/v1/chat/completions` (403-gated on `API_SERVER_KEY`, and the client can resume the session by sending the header again);
- native `/api/sessions/{session_id}/chat[/stream]` routes (the id is in the request path);
- `/v1/runs` (body `session_id`, response-chain id, declared session key, or the `run_id` fallback — all held by the client, and a wake-injected turn lands in session history a later run on the same id loads).

Everything else (fingerprint-derived ids, `/v1/responses` fresh-uuid sessions, undeclared binders) keeps the forced-synchronous fallback with its honest note. The flag is captured in `_capture_origin()` **before** child construction (same clobber-proof timing as `origin_wake_sid`), read fail-closed, and never falls back to `os.environ` — it deliberately stays out of `_VAR_MAP` so a leaked env var cannot grant wake authority.

**Credit:** #98619 is a first-hand production report by @shojikumaru, who — with Alpha — did the original investigation and independently verified the original head's before/after behavior on a real isolated `api_server` setup (see their comment on this PR). This rebuild stands on that work.

## Related Issue

Fixes #98619

## Type of Change

- [x] 🔒 Security fix (fails closed a delivery-authority boundary; issue carries `sweeper:risk-security-boundary`)

## Changes Made

- `gateway/session_context.py` — new `_SESSION_WAKE_CAPABLE` ContextVar (default `_UNSET`, deliberately outside `_VAR_MAP`: no env fallback, no subprocess-bridge export) + `set_session_vars(wake_capable=None|"1"|"")` binding + `wake_capable_session()` reader (True only for literal `"1"`); `clear_session_vars`/`reset_session_vars` reset it to `_UNSET`.
- `gateway/platforms/api_server.py` — `_bind_api_server_session(..., wake_capable="")` default-denies at the single chokepoint; `_run_agent(..., wake_capable="")` threads it; `_prepare_session_chat` opts the native `/api/sessions/{id}` routes in with `"1"`.
- `gateway/platforms/api_server_openai_routes.py` — `/v1/chat/completions` passes `wake_capable=("1" if provided_session_id else "")`; streaming inherits via `run_kwargs`.
- `gateway/platforms/api_server_runs.py` — `/v1/runs` binds with `wake_capable="1"` (audited: client-held ids).
- `tools/delegate_tool_dispatch.py` — `_capture_origin()` also captures the capability pre-child-construction (read error ⇒ `False`); `_Batch.origin_wake_capable`; `_resolve_async_wake_sid()` requires **both** the bound id and the declared capability, with an honest "bound but not wake-capable → synchronous" log line.
- `tests/tools/test_delegate_apiserver_background.py` — public-omission regression (a binder that never declares the flag stays synchronous), derived-id regression, `wake_capable_session()` default-deny/env-no-fallback unit test, positive control updated to declare `"1"`.
- `tests/gateway/test_api_server.py` — route-level provenance test: explicit header ⇒ `"1"`, header-less ⇒ `""`.

## How to Test

1. `pytest tests/tools/test_delegate_apiserver_background.py -q` → **5 passed** (positive control dispatches background only with a declared `"1"`; omission and derived ids stay synchronous).
2. `pytest "tests/gateway/test_api_server.py::TestSessionIdHeader" -q` → **3 passed** (route threads `"1"`/`""` by header provenance).
3. Regression sweep on the touched surfaces: `pytest tests/gateway/test_api_server.py tests/gateway/test_api_server_runs.py tests/gateway/test_session_context_inheritance.py tests/gateway/test_wake_delivery.py tests/tools/test_delegate_batch_tag.py tests/tools/test_delegate_batch_validation.py tests/tools/test_delegate_capability_inheritance.py tests/tools/test_delegate_control_actions.py tests/tools/test_delegate_composite_toolsets.py tests/tools/test_async_delegation.py -q` → **all passed** (292 relevant tests green total).
4. Observed result: a header-less client's `delegate_task(background=true)` now runs synchronously and returns the result in the same response — exactly the "After" trace @shojikumaru verified on the original head — while explicit-header/native/`/v1/runs` sessions keep detached dispatch.

### Code

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — or N/A (behavior documented in docstrings at the chokepoint and the ContextVar)

### DIFF
--- a/gateway/platforms/api_server_openai_routes.py
+++ b/gateway/platforms/api_server_openai_routes.py
@@ -473,6 +473,13 @@ class OpenAICompatRoutesMixin:
             try:
                 db = await self._ensure_session_db_async()
                 if db is not None:
+                    # #98619/#13437: a client-addressed id from before a compression rotation
+                    # must adopt the live continuation tip — history loads from it, the turn and
+                    # the wake target bind it, and a detached delegation delivery row persisted
+                    # on the tip is what this continuation consumes. Same canonical resolution
+                    # the delivery writer (gateway/wake.py) and /v1/runs use; fails open.
+                    from gateway.platforms.api_server_runs import _resolve_live_session_id
+                    session_id = await _resolve_live_session_id(self, provided_session_id)
                     history = await asyncio.to_thread(db.get_messages_as_conversation, session_id)
             except Exception as e:
                 logger.warning("Failed to load session history for %s: %s", session_id, e)
@@ -530,9 +537,14 @@ class OpenAICompatRoutesMixin:
             agent_task, agent_ref = self._spawn_stream_agent(
                 _stream_q, tool_start_callback=_on_tool_start,
                 tool_complete_callback=_on_tool_complete, **run_kwargs)
+            # #13437 identity contract: an explicit-header client keeps addressing the id it
+            # sent; the response echoes that stable id while reads/writes adopt the live tip,
+            # so a rotation mid-turn (after these headers are prepared) never changes what the
+            # client should send next — it re-sends the same id and the tip resolution above
+            # finds whatever session is live by then.
             return await self._write_sse_chat_completion(
                 request, completion_id, model_name, created, _stream_q,
-                agent_task, agent_ref, session_id=session_id,
+                agent_task, agent_ref, session_id=(provided_session_id or session_id),
                 gateway_session_key=gateway_session_key)
 
         async def _compute_completion():
@@ -549,7 +561,10 @@ class OpenAICompatRoutesMixin:
         if err_msg:
             err_msg = _redact_api_error_text(err_msg)
         finish_reason = _finish_reason(completed, is_partial, is_failed, err_msg)
-        response_headers = {"X-Hermes-Session-Id": result.get("session_id", session_id)}
+        # Same #13437 identity contract as the SSE path: an explicit-header client is echoed
+        # the stable id it sent; a fingerprint-derived (header-less) turn keeps reporting the
+        # id the agent actually resolved, so headerless clients still learn where the turn went.
+        response_headers = {"X-Hermes-Session-Id": (provided_session_id or result.get("session_id", session_id))}
         if gateway_session_key:
             response_headers["X-Hermes-Session-Key"] = gateway_session_key
         # Hard fail (no usable text AND a real failure) -> 502 OpenAI error envelope so SDK
--- a/gateway/platforms/api_server_openai_routes.py
+++ b/gateway/platforms/api_server_openai_routes.py
@@ -494,7 +494,13 @@ class OpenAICompatRoutesMixin:
         run_kwargs = dict(
             user_message=user_message, conversation_history=history,
             ephemeral_system_prompt=system_prompt, session_id=session_id,
-            gateway_session_key=gateway_session_key, **agent_overrides, route=route)
+            gateway_session_key=gateway_session_key, **agent_overrides, route=route,
+            # #98619: only an explicitly provided X-Hermes-Session-Id is wake-capable (the
+            # header is 403-gated on API_SERVER_KEY, so the wake self-post can authenticate
+            # and the client can resume the session by sending it again). A fingerprint-derived
+            # id from a header-less client is NOT: delegate_task keeps its forced-sync fallback
+            # there — the wake would hard-fail or land in history that client never reloads.
+            wake_capable=("1" if provided_session_id else ""))
         if stream:
             _stream_q = ThreadSafeAsyncQueue()
             # tool_call_ids with an emitted "running": a "completed" without one (internal/
--- a/gateway/platforms/api_server.py
+++ b/gateway/platforms/api_server.py
@@ -3035,7 +3035,11 @@ class APIServerAdapter(OpenAICompatRoutesMixin, BasePlatformAdapter):
             gateway_session_key=gateway_session_key, route=route, session_model=session_model,
             requested_runtime=runtime_request.get("requested") or {},
             route_source=runtime_request.get("route_source") or "global",
-            confirmed_runtime_lock=lock_active, **agent_overrides)
+            confirmed_runtime_lock=lock_active,
+            # #98619: the client addresses this session by construction — the id is in the
+            # request path (/api/sessions/{session_id}/chat) — so a wake self-post lands where
+            # the client will read it. The audited native-session opt-in.
+            wake_capable="1", **agent_overrides)
         return {
             "gateway_session_key": gateway_session_key, "session_id": session_id, "body": body,
             "user_message": user_message, "runtime_request": runtime_request,
@@ -3542,11 +3546,19 @@ class APIServerAdapter(OpenAICompatRoutesMixin, BasePlatformAdapter):
     @staticmethod
     def _bind_api_server_session(
         *, chat_id: str = "", session_key: str = "", session_id: str = "",
-        browser_control_principal: str = "", browser_control_transport_family: str = "") -> list:
+        browser_control_principal: str = "", browser_control_transport_family: str = "",
+        wake_capable: str = "") -> list:
         """Bind session contextvars for an API-server agent run — the SINGLE chokepoint for every
         agent-entry path. Hardwires ``platform="api_server"`` + ``async_delivery=False`` (HTTP
         can never wake the agent after the turn) so no route reintroduces the silent no-op bug.
         Returns reset tokens for ``clear_session_vars`` in a ``finally`` (request-scoped).
+
+        ``wake_capable`` is the separate #98619 gate and DEFAULT-DENIES here: only an audited
+        producer whose client can address the bound id again (explicit X-Hermes-Session-Id —
+        403-gated on API_SERVER_KEY — a native /api/sessions/{id} route, /v1/runs) passes "1";
+        a fingerprint-derived id (header-less OpenAI-compatible client) stays "" and keeps
+        delegate_task's forced-sync fallback — the wake self-post could never deliver where
+        that client reads. A route that says nothing grants no wake authority.
 
         See #10760.
         """
@@ -3555,7 +3567,7 @@ class APIServerAdapter(OpenAICompatRoutesMixin, BasePlatformAdapter):
             platform="api_server", chat_id=chat_id, session_key=session_key, session_id=session_id,
             browser_control_principal=browser_control_principal,
             browser_control_transport_family=browser_control_transport_family,
-            async_delivery=False, cron_session="")
+            async_delivery=False, cron_session="", wake_capable=wake_capable)
 
     def _turn_runtime_metadata(
         self, agent: Any, *, route: Optional[Dict[str, Any]], requested_runtime: Optional[Dict[str, Any]],
@@ -3629,11 +3641,15 @@ class APIServerAdapter(OpenAICompatRoutesMixin, BasePlatformAdapter):
         requested_provider: Optional[str] = None, model_options: Optional[Dict[str, Any]] = None,
         route: Optional[Dict[str, Any]] = None, session_model: Optional[str] = None,
         requested_runtime: Optional[Dict[str, Any]] = None, route_source: str = "global",
-        confirmed_runtime_lock: bool = False, bind_declared_conversation: bool = False) -> tuple:
+        confirmed_runtime_lock: bool = False, bind_declared_conversation: bool = False,
+        wake_capable: str = "") -> tuple:
         """Create an agent and run one turn in a thread executor -> ``(result, usage)``.
         ``agent_ref[0]`` receives the agent so SSE writers can interrupt it; ``active_run_id``
         registers it in ``_active_run_agents``. Under a confirmed model lock the actual
-        provider/model must match or the turn fails; ``runtime`` metadata is attached."""
+        provider/model must match or the turn fails; ``runtime`` metadata is attached.
+        ``wake_capable`` declares #98619 session-id provenance and default-denies: only audited
+        producers whose client can address the id again pass "1" (see
+        ``_bind_api_server_session``)."""
         loop = asyncio.get_running_loop()
         # ContextVars do not follow run_in_executor threads: capture here, re-enter in _run().
         request_profile = _api_request_profile.get()
@@ -3647,7 +3663,8 @@ class APIServerAdapter(OpenAICompatRoutesMixin, BasePlatformAdapter):
                     chat_id=session_id or "", session_key=gateway_session_key or session_id or "",
                     session_id=session_id or "",
                     browser_control_principal=request_browser_control_principal,
-                    browser_control_transport_family=request_browser_control_transport_family)
+                    browser_control_transport_family=request_browser_control_transport_family,
+                    wake_capable=wake_capable)
                 agent = None
                 try:
                     agent = self._create_agent(
--- a/gateway/platforms/api_server_runs.py
+++ b/gateway/platforms/api_server_runs.py
@@ -354,6 +354,23 @@ def _drop_run_transport(self, run_id: str) -> None:
     _forget_run(self, run_id, self._run_streams, self._run_streams_created)
 
 
+async def _resolve_live_session_id(self, session_id: str) -> str:
+    """Adopt the live compression-continuation tip for a client-addressed session (#98619):
+    a /v1/runs run bound to a pre-rotation id would otherwise load a stale history slice and
+    write its turn into the closed parent (CompressionSessionClosedError). Same canonical
+    resolution ``/api/sessions/{id}/messages`` reads use; fails open to the original id."""
+    db = await self._ensure_session_db_async()
+    resolver = getattr(db, "resolve_resume_session_id", None) if db is not None else None
+    if not callable(resolver):
+        return session_id
+    try:
+        resolved = await asyncio.to_thread(resolver, session_id)
+        return str(resolved) if resolved else session_id
+    except Exception:
+        logger.debug("/v1/runs live-session resolve failed for %s", session_id, exc_info=True)
+        return session_id
+
+
 async def _handle_runs(self, request: "web.Request", *, _api_server) -> "web.Response":
     """POST /v1/runs — start an agent run, return run_id immediately."""
     _openai_error = _api_server._openai_error
@@ -427,6 +444,11 @@ async def _handle_runs(self, request: "web.Request", *, _api_server) -> "web.Res
     _declared_selected = not session_id and bool(gateway_session_key)
     selected_session_id = session_id or (
         self._declared_conversation_session(gateway_session_key) if _declared_selected else None)
+    # A client-addressed id from before a compression rotation must adopt the live tip (#98619):
+    # history loads from it, the turn writes to it, and a detached delivery row persisted to it
+    # is what the next same-id run consumes below.
+    if selected_session_id:
+        selected_session_id = await _resolve_live_session_id(self, str(selected_session_id))
     session_id = selected_session_id or run_id
     # History loads for the session the request actually selected — including one resolved from
     # a declared X-Hermes-Session-Key, whose persisted delivery rows must reach the next
--- a/gateway/platforms/api_server_runs.py
+++ b/gateway/platforms/api_server_runs.py
@@ -486,7 +486,12 @@ def _run_agent_sync(self, run: _RunLaunch, agent, approval_notify, *, _api_serve
             session_tokens = self._bind_api_server_session(
                 chat_id=session_id or "", session_key=run.approval_session_key, session_id=session_id or "",
                 browser_control_principal=run.browser_control_principal,
-                browser_control_transport_family=run.browser_control_transport_family)
+                browser_control_transport_family=run.browser_control_transport_family,
+                # #98619 audited opt-in: every /v1/runs session id is one the client can address
+                # again (body session_id, a response chain id, a declared session key, or the
+                # run_id fallback returned by the create response) — and a wake-injected turn
+                # lands in session history a later run on the same id will load.
+                wake_capable="1")
             if session_tokens:
                 resets.append((session_tokens, clear_session_vars))
             if run.agent_kwargs["room_dispatch"] is not None:
--- a/gateway/platforms/api_server_runs.py
+++ b/gateway/platforms/api_server_runs.py
@@ -319,7 +319,9 @@ class _RunLaunch:
     user_message: str
     conversation_history: List[Dict[str, str]]
     # #98619: only continuation paths that reload session history may grant wake authority —
-    # a previous_response_id continuation consumes its ResponseStore snapshot instead.
+    # a previous_response_id continuation consumes its ResponseStore snapshot instead, and a
+    # caller-supplied conversation_history is authoritative for the turn; neither consumes a
+    # SessionDB delivery row, so both stay default-denied.
     wake_capable: bool
     agent_kwargs: dict  # ``_create_agent`` keyword arguments (prompt, model overrides, route, room policy)
     request_profile: Any
@@ -455,7 +457,10 @@ async def _handle_runs(self, request: "web.Request", *, _api_server) -> "web.Res
     # same-key run's context (#98619).  previous_response_id continuations keep their
     # ResponseStore snapshot as history (they cannot consume a SessionDB delivery row and are
     # accordingly denied wake capability in _run_agent_sync); the fresh run_id fallback has
-    # nothing persisted to load yet.
+    # nothing persisted to load yet.  Wake authority is fixed here, before the load can
+    # overwrite ``conversation_history``: a caller-supplied history is authoritative for this
+    # turn, never consumes the SessionDB delivery row, and is denied on the same contract.
+    wake_capable = not previous_response_id and not conversation_history
     if not conversation_history and selected_session_id and not previous_response_id:
         conversation_history = await self._conversation_history_for_session(str(selected_session_id))
     q = self._run_streams[run_id] = asyncio.Queue()
@@ -476,7 +481,7 @@ async def _handle_runs(self, request: "web.Request", *, _api_server) -> "web.Res
         self._run_idempotency_ids.add(run_id)
     launch = _RunLaunch(
         self, run_id, q, session_id, gateway_session_key, _declared_selected, user_message,
-        conversation_history, not previous_response_id,
+        conversation_history, wake_capable,
         agent_kwargs=dict(
             ephemeral_system_prompt=instructions, session_id=session_id, gateway_session_key=gateway_session_key,
             route=route, room_dispatch=room_dispatch, room_execution_policy=room_execution_policy,
@@ -526,7 +531,9 @@ def _run_agent_sync(self, run: _RunLaunch, agent, approval_notify, *, _api_serve
                 # SessionDB in _handle_runs), or the run_id fallback the client can post back
                 # as body.session_id.  A previous_response_id continuation consumes its
                 # ResponseStore snapshot instead and can never see a SessionDB delivery row,
-                # so it stays default-denied until a merge contract exists for that chain.
+                # so it stays default-denied until a merge contract exists for that chain;
+                # likewise a caller-supplied conversation_history is authoritative for the
+                # turn and never reads the delivery row, so it is denied the same way.
                 wake_capable="1" if run.wake_capable else "")
             if session_tokens:
                 resets.append((session_tokens, clear_session_vars))
--- a/gateway/platforms/api_server_runs.py
+++ b/gateway/platforms/api_server_runs.py
@@ -318,6 +318,9 @@ class _RunLaunch:
     declared_selected: bool
     user_message: str
     conversation_history: List[Dict[str, str]]
+    # #98619: only continuation paths that reload session history may grant wake authority —
+    # a previous_response_id continuation consumes its ResponseStore snapshot instead.
+    wake_capable: bool
     agent_kwargs: dict  # ``_create_agent`` keyword arguments (prompt, model overrides, route, room policy)
     request_profile: Any
     browser_control_principal: Any
@@ -416,15 +419,23 @@ async def _handle_runs(self, request: "web.Request", *, _api_server) -> "web.Res
     limited = self._concurrency_limited_response()
     if limited is not None:
         return limited
-    if not conversation_history and session_id and not previous_response_id:
-        conversation_history = await self._conversation_history_for_session(str(session_id))
     run_id = f"run_{uuid.uuid4().hex}"
     self._run_owners[run_id] = self._run_idempotency_scope(request)
     # Same precedence as /v1/responses: body session_id > response chain > X-Hermes-Session-Key
     # conversation > run_id (which would otherwise re-key every affinity surface per run).
     # An explicit or chained session owns its routing key and is never rebound to the header.
     _declared_selected = not session_id and bool(gateway_session_key)
-    session_id = session_id or self._declared_conversation_session(gateway_session_key) or run_id
+    selected_session_id = session_id or (
+        self._declared_conversation_session(gateway_session_key) if _declared_selected else None)
+    session_id = selected_session_id or run_id
+    # History loads for the session the request actually selected — including one resolved from
+    # a declared X-Hermes-Session-Key, whose persisted delivery rows must reach the next
+    # same-key run's context (#98619).  previous_response_id continuations keep their
+    # ResponseStore snapshot as history (they cannot consume a SessionDB delivery row and are
+    # accordingly denied wake capability in _run_agent_sync); the fresh run_id fallback has
+    # nothing persisted to load yet.
+    if not conversation_history and selected_session_id and not previous_response_id:
+        conversation_history = await self._conversation_history_for_session(str(selected_session_id))
     q = self._run_streams[run_id] = asyncio.Queue()
     created_at = self._run_streams_created[run_id] = time.time()
     self._run_approval_sessions[run_id] = run_id  # approval session key (see _RunLaunch)
@@ -443,7 +454,7 @@ async def _handle_runs(self, request: "web.Request", *, _api_server) -> "web.Res
         self._run_idempotency_ids.add(run_id)
     launch = _RunLaunch(
         self, run_id, q, session_id, gateway_session_key, _declared_selected, user_message,
-        conversation_history,
+        conversation_history, not previous_response_id,
         agent_kwargs=dict(
             ephemeral_system_prompt=instructions, session_id=session_id, gateway_session_key=gateway_session_key,
             route=route, room_dispatch=room_dispatch, room_execution_policy=room_execution_policy,
@@ -487,11 +498,14 @@ def _run_agent_sync(self, run: _RunLaunch, agent, approval_notify, *, _api_serve
                 chat_id=session_id or "", session_key=run.approval_session_key, session_id=session_id or "",
                 browser_control_principal=run.browser_control_principal,
                 browser_control_transport_family=run.browser_control_transport_family,
-                # #98619 audited opt-in: every /v1/runs session id is one the client can address
-                # again (body session_id, a response chain id, a declared session key, or the
-                # run_id fallback returned by the create response) — and a wake-injected turn
-                # lands in session history a later run on the same id will load.
-                wake_capable="1")
+                # #98619 audited opt-in: the /v1/runs session id is wake-capable only when its
+                # own continuation path reloads session history — an explicit body/chained
+                # session id or a declared X-Hermes-Session-Key conversation (both load
+                # SessionDB in _handle_runs), or the run_id fallback the client can post back
+                # as body.session_id.  A previous_response_id continuation consumes its
+                # ResponseStore snapshot instead and can never see a SessionDB delivery row,
+                # so it stays default-denied until a merge contract exists for that chain.
+                wake_capable="1" if run.wake_capable else "")
             if session_tokens:
                 resets.append((session_tokens, clear_session_vars))
             if run.agent_kwargs["room_dispatch"] is not None:
--- a/gateway/session_context.py
+++ b/gateway/session_context.py
@@ -52,6 +52,18 @@ _SESSION_VARS = (
 # adapters (API server, Kanban workers) opt OUT via ``supports_async_delivery = False`` at bind.
 _SESSION_ASYNC_DELIVERY = ContextVar("HERMES_SESSION_ASYNC_DELIVERY", default=_UNSET)
 
+# Whether the bound api_server chat id is one its CLIENT can address again — the precondition for
+# gateway.wake's /v1/chat/completions self-post to deliver a background delegation anywhere the
+# requester will read (#98619).  "1" = wake-capable, declared by an audited producer whose client
+# holds or can resume the id (explicit X-Hermes-Session-Id, a native /api/sessions/{id} id, a
+# /v1/runs id); "" = declared NOT wake-capable (a fingerprint-derived id from a header-less
+# OpenAI-compatible client, which that client never reloads).  _UNSET = the binding never declared
+# it.  Unlike async delivery above, _UNSET FAILS CLOSED (see ``wake_capable_session``): wake
+# authority is proof-carrying, so a binder that omits it must not silently acquire it.  Deliberately
+# NOT in ``_VAR_MAP``: no ``os.environ`` fallback (a leaked env var must not grant wake authority)
+# and no subprocess-env-bridge export (children re-derive provenance from their own binding).
+_SESSION_WAKE_CAPABLE = ContextVar("HERMES_SESSION_WAKE_CAPABLE", default=_UNSET)
+
 # Cron auto-delivery vars, set per-job in run_job() so concurrent jobs don't clobber.
 _CRON_AUTO_DELIVER_PLATFORM = ContextVar("HERMES_CRON_AUTO_DELIVER_PLATFORM", default=_UNSET)
 _CRON_AUTO_DELIVER_CHAT_ID = ContextVar("HERMES_CRON_AUTO_DELIVER_CHAT_ID", default=_UNSET)
@@ -107,11 +119,17 @@ def set_session_vars(
     user_name: str = "", scope_id: str = "", session_key: str = "", session_id: str = "",
     message_id: str = "", profile: str = "", browser_control_principal: str = "",
     browser_control_transport_family: str = "", cwd: str = "", async_delivery: bool = True,
-    ui_session_id: str = "", cron_session: Any = _UNSET,
+    ui_session_id: str = "", cron_session: Any = _UNSET, wake_capable: str | None = None,
 ) -> list:
     """Set all session context variables and return reset tokens.  Call
     ``clear_session_vars(tokens)`` in a ``finally``; not nestable, clearing resets every var
-    to ``""`` rather than restoring prior values (tokens are accepted only for API compat)."""
+    to ``""`` rather than restoring prior values (tokens are accepted only for API compat).
+
+    ``wake_capable`` declares whether the bound chat id is one the client can address again:
+    ``"1"`` (audited producers — explicit session-id header, native API sessions, /v1/runs) or
+    ``""`` / omitted (default-deny, #98619).  ``None`` leaves the var at ``_UNSET`` ("never
+    declared"), which ``wake_capable_session()`` treats as NOT capable — an omitted declaration
+    cannot grant wake authority."""
     global _session_context_engaged
     _session_context_engaged = True
     values = (
@@ -121,6 +139,7 @@ def set_session_vars(
     )
     tokens = [var.set(value) for var, value in zip(_SESSION_VARS, values)]
     tokens.append(_SESSION_ASYNC_DELIVERY.set(bool(async_delivery)))
+    tokens.append(_SESSION_WAKE_CAPABLE.set(_UNSET if wake_capable is None else wake_capable))
     _runtime_cwd("set_session_cwd", cwd)
     return tokens
 
@@ -128,10 +147,13 @@ def set_session_vars(
 def clear_session_vars(tokens: list) -> None:
     """Mark session context variables as explicitly cleared (``""``, not ``_UNSET``), so
     ``get_session_env`` returns empty instead of stale ``os.environ`` values.  Async-delivery
-    goes back to ``_UNSET``: a cleared context is default-supported, not opted-out."""
+    goes back to ``_UNSET``: a cleared context is default-supported, not opted-out.  Wake
+    capability goes back to ``_UNSET`` too — but for the opposite reason: a cleared context has
+    declared nothing, and an undeclared capability FAILS CLOSED (#98619)."""
     for var in _SESSION_VARS:
         var.set("")
     _SESSION_ASYNC_DELIVERY.set(_UNSET)
+    _SESSION_WAKE_CAPABLE.set(_UNSET)
     _runtime_cwd("clear_session_cwd")
 
 
@@ -139,10 +161,12 @@ def reset_session_vars() -> None:
     """Reset every session var to ``_UNSET`` ("never bound here") for THIS context.  Call at
     the top of a fresh task *before* it binds: ``create_task`` snapshots the context, so B's
     task inherits A's already-set vars and a subprocess spawned before B binds would read A's
-    identity.  ``_SESSION_ASYNC_DELIVERY`` (outside ``_VAR_MAP``) is reset explicitly too."""
+    identity.  ``_SESSION_ASYNC_DELIVERY`` and ``_SESSION_WAKE_CAPABLE`` (outside ``_VAR_MAP``)
+    are reset explicitly too."""
     for var in _VAR_MAP.values():
         var.set(_UNSET)
     _SESSION_ASYNC_DELIVERY.set(_UNSET)
+    _SESSION_WAKE_CAPABLE.set(_UNSET)
     _runtime_cwd("clear_session_cwd")
 
 
@@ -191,3 +215,12 @@ def async_delivery_supported() -> bool:
         return False
     value = _SESSION_ASYNC_DELIVERY.get()
     return True if value is _UNSET else bool(value)
+
+
+def wake_capable_session() -> bool:
+    """Whether the current session's bound api_server chat id was explicitly declared resumable
+    by its client (#98619) — True only for the literal ``"1"``.  Default-deny: ``_UNSET`` (the
+    binding never declared it) and ``""`` (declared not capable) both count as NOT wake-capable.
+    Never falls back to ``os.environ`` — wake authority is proof-carrying and cannot leak in
+    from the environment."""
+    return _SESSION_WAKE_CAPABLE.get() == "1"
--- a/gateway/wake.py
+++ b/gateway/wake.py
@@ -92,6 +92,20 @@ async def persist_delegation_delivery(adapter: Any, *, text: str, session_id: st
     if db is None:
         raise RuntimeError("persist_delegation_delivery: api_server SessionDB unavailable — "
                            f"cannot persist completion for session {session_id}")
+    # #98619: the parent run may have compressed/rotated between dispatch and this detached
+    # completion — the captured origin id is then a closed parent and the append below is
+    # rejected with CompressionSessionClosedError forever (the watcher retries the same stale
+    # id). Adopt the live continuation tip first, the same canonical resolution
+    # /api/sessions/{id}/messages reads use, so the delivery row lands where the next run and
+    # the messages endpoint both resolve. Fails open to the original id.
+    resolver = getattr(db, "resolve_resume_session_id", None)
+    if callable(resolver):
+        try:
+            resolved = await asyncio.to_thread(resolver, session_id)
+            if resolved:
+                session_id = str(resolved)
+        except Exception:
+            logger.debug("delegation delivery continuation resolve failed for %s", session_id, exc_info=True)
     await asyncio.to_thread(
         db.append_message, session_id, "user", content=text,
         display_kind="async_delegation_complete", display_metadata=_delegation_display_metadata(evt or {}),
--- a/tests/gateway/test_api_server.py
+++ b/tests/gateway/test_api_server.py
@@ -2385,6 +2385,40 @@ class TestSessionIdHeader:
             assert call_kwargs["conversation_history"] == db_history
             assert call_kwargs["user_message"] == "new question"
 
+    @pytest.mark.asyncio
+    async def test_wake_capability_follows_session_id_provenance(self, auth_adapter):
+        """#98619: wake_capable must be "1" only when the session id was
+        explicitly provided (a header client that can resume it), "" when it is
+        fingerprint-derived (a header-less client) — delegate_task's background
+        gate keys on it to keep the forced-sync fallback where the wake
+        self-post could never deliver."""
+        mock_result = {"final_response": "OK", "messages": [], "api_calls": 1}
+        app = _create_app(auth_adapter)
+        async with TestClient(TestServer(app)) as cli:
+            with patch.object(auth_adapter, "_run_agent", new_callable=AsyncMock) as mock_run:
+                mock_run.return_value = (mock_result, {"input_tokens": 0, "output_tokens": 0, "total_tokens": 0})
+
+                resp = await cli.post(
+                    "/v1/chat/completions",
+                    headers={"X-Hermes-Session-Id": "client-held-session", "Authorization": "Bearer sk-secret"},
+                    json={"model": "hermes-agent", "messages": [{"role": "user", "content": "hi"}]},
+                )
+                assert resp.status == 200
+                assert mock_run.call_args.kwargs["wake_capable"] == "1"
+
+            with patch.object(auth_adapter, "_run_agent", new_callable=AsyncMock) as mock_run:
+                mock_run.return_value = (mock_result, {"input_tokens": 0, "output_tokens": 0, "total_tokens": 0})
+
+                # Header-less: the session id is derived from the message
+                # fingerprint — this client will never resume that session.
+                resp = await cli.post(
+                    "/v1/chat/completions",
+                    headers={"Authorization": "Bearer sk-secret"},
+                    json={"model": "hermes-agent", "messages": [{"role": "user", "content": "hi"}]},
+                )
+                assert resp.status == 200
+                assert mock_run.call_args.kwargs["wake_capable"] == ""
+
 
 # ---------------------------------------------------------------------------
 # X-Hermes-Session-Key header (long-term memory scoping)
--- a/tests/gateway/test_api_server.py
+++ b/tests/gateway/test_api_server.py
@@ -2359,6 +2359,8 @@ class TestSessionIdHeader:
         ]
         mock_db = MagicMock()
         mock_db.get_messages_as_conversation.return_value = db_history
+        # Non-rotated control: the canonical tip resolver resolves the id to itself.
+        mock_db.resolve_resume_session_id.side_effect = lambda sid: sid
         auth_adapter._session_db = mock_db
         app = _create_app(auth_adapter)
         async with TestClient(TestServer(app)) as cli:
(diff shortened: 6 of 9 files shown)


andrexibiza (commented):
I like the core diagnosis here. The important correction is not “API server has a session id”; it is “this exact request has a return address the originating client can actually resume.” Capturing that provenance before child construction and preserving the existing synchronous fallback is the right direction for #98619.

I reviewed exact head `09e842ef6cef5ea25efb6d4ab0ced54f4d4cfc5f`, the full five-file diff, both streaming/non-streaming API-server call sites, the central session-context lifetime, the background delegation wake gate, the new regressions, #98619, current-main ancestry, exact-head CI, and the adjacent async-delivery work. I found two landing blockers and one provenance/credit gate.

### P1 — the new provenance primitive is still default-allow at the central binding boundary

`gateway/session_context.py:256` adds:

```py
wake_capable: str = "1"
```

and `set_session_vars()` always commits that value into `_SESSION_WAKE_CAPABLE`. That contradicts the contract immediately above it, which says `_UNSET` means the binding never declared wake capability and readers should fail closed.

This matters beyond the two API-server call sites fixed in this PR. A capability/provenance bit only closes the defect class if omission cannot grant it. Right now any existing or future binder that calls `set_session_vars(...)` without knowing about `wake_capable` silently acquires wake authority. The central helper therefore still turns “no proof supplied” into “allowed.” That is the same boundary failure in a new representation.

The new regression exposes this mechanically. `test_apiserver_wake_capability_unbound_fails_closed` first calls `set_session_vars(...)` *without* `wake_capable`, but that public call has already set the flag to `"1"`. The test then has to reach into private implementation state and manually undo it:

```py
sc._VAR_MAP["HERMES_SESSION_WAKE_CAPABLE"].set(sc._UNSET)
```

So the test proves the consumer behaves correctly if an `_UNSET` magically exists; it does not prove that a real binder which forgets to declare the capability fails closed. In production, that forgotten binder gets `"1"`.

**Required fix:** make the central binding API default to non-authorized (`""` or `_UNSET`), then explicitly opt in every audited producer that really owns a resumable return address. Replace the private-ContextVar surgery regression with a real public binding/caller omission witness: bind a session without declaring wake capability and prove `delegate_task(background=True)` stays synchronous. Keep positive controls for explicit `X-Hermes-Session-Id`, native API sessions, `/v1/runs`, and the normal messaging-gateway surfaces that truly support post-turn delivery.

That makes the new bit proof-carrying rather than an ambient compatibility assumption.

### P1 — this adds routing authority to two live godfiles instead of extracting the owner

The changed `gateway/platforms/api_server.py` hunks are already around line 7,200, and the changed `tools/delegate_tool.py` wake-policy hunk is around line 4,200. Both are well past the repository's 2K owner ceiling. This PR is adding a new authority/provenance contract, not a leaf cosmetic change, so growing those files further leaves the ownership boundary worse even if the local logic is correct.

Please extract the wake-addressability/provenance decision into a bounded owner and leave these surfaces as thin adapters: API-server produces the typed/current capability; delegation consumes it. The tests should target that seam directly plus one real API-server → delegate integration witness. This is also important because `gateway/platforms/api_server.py` currently collides with open #81806, while `tools/delegate_tool.py` collides with open #98348. The second landing in each pair must recompose through the surviving owner rather than restoring another branch's embedded policy.

### Provenance / contributor credit gate

#98619 is not an anonymous specification. The issue is a first-hand production report by `shojikumaru` and explicitly records the Sho + Alpha investigation. The PR currently has only `Fixes #98619`; the commit likewise names the issue but not the prior contributors. Please preserve that lineage explicitly in the PR/commit/closure surfaces rather than allowing the implementation to erase the source of the defect and reproduction.

### Interlock / merge-order map

These are related, but not duplicates:

- **#96086** and **#82704** are complementary durable-delivery ownership fixes in `gateway/run.py`: they stop a gateway from consuming delivery attempts when it cannot actually route a completion. #98630 sits earlier in the pipeline and decides whether Hermes may promise detached delivery at all for this request.
- **#81806** is adjacent API-server async-session/drain observability and directly overlaps `gateway/platforms/api_server.py`. Preserve its session/drain semantics if it lands first; rebase this producer onto that owner rather than replaying the old file.
- **#98348** is adjacent delegated-child lifetime ownership and directly overlaps `tools/delegate_tool.py`. Its timeout/retirement semantics are independent of this wake-capability gate and must survive either merge order.
- The wider class also touches the same routing/provenance theorem as the session-binding work: a session coordinate can identify a candidate destination, but it cannot by itself prove that a post-turn effect is deliverable to the requester.

### Exact-object acceptance

The submitted one-commit head is well tested: exact-head CI `33317442809`, Docker `33317442257`, and Nix `33317442216` are green; CI includes Python/e2e, Windows/macOS, blocking footguns/ruff, contributor attribution, supply-chain checks, and `All required checks pass`. That is good evidence for this exact submitted object.

However, the branch is currently `1 ahead / 3 behind` current `main@5cc1369fa298021f8c740de154ff8c37c30bdcc8`. I checked the three intervening main commits; they are skill-tooling changes and do not overlap these five paths, so this should be a straightforward recomposition, but the landing object still needs to be rebuilt on current main and receive fresh exact-head receipts.

The useful part here is strong: separating “session id exists” from “the client can consume a later wake” is exactly the right correction, and the forced-sync fallback keeps correctness ahead of latency. Make the capability impossible to acquire by omission, put its policy behind a bounded owner, preserve the reporter lineage, and this can close the whole shape rather than only the observed API-server branch. 🚀

shojikumaru (commented):
**Real-setup verification — tested at head `09e842ef6c` (this PR's current head) vs. its merge-base `dce2ecb8a9`**

I'm the reporter of #98619; this is an independent run of the PR on a real `api_server` instance, plus the two-sided test run, plus a note on where the gate has moved on current `main` (the branch is conflicting).

**Setup:** isolated checkout + venv on Ubuntu 24.04 (Python 3.12.3), own `HERMES_HOME`, `api_server` on `127.0.0.1:18642` with a strong `API_SERVER_KEY` (so this is the *authenticated* case ② from the issue — the current startup guard refuses to run without a key, so case ① "no key → wake RuntimeError" is not reachable on this build). Model: local Ollama (`qwen3:8b` re-created with `num_ctx 65536`) — no external provider, no production credentials. Client: plain `curl` to `POST /v1/chat/completions` with `Authorization` only, **no `X-Hermes-Session-Id`**, asking the agent to call `delegate_task(background=true)` once with the goal "reply with the single word PONG". Identical request on both sides; the derived session id was the same on both (`api-546cc1dbe338afb4`), as expected for a fingerprint.

**Before (base `dce2ecb8a9`):** the result never reaches the client.

```
gateway log:
22:30:18 INFO tools.delegate_tool: delegate_task: async delivery unsupported on this session, but a session id is bound (api-546cc1dbe338afb4) — dispatching in the background and waking the session via self-post when it completes instead of forcing synchronous execution.
22:30:18 INFO tools.async_delegation: Dispatched async delegation batch deleg_23b75e24 (1 task(s), session_key=api-54...fb4)
22:30:18 INFO agent.tool_executor: tool delegate_task completed (0.51s, 983 chars)

HTTP 200 to the header-less client (after 477 s):
"content": "The delegated task is running in the background and will send its result as a new message when complete."

…then, ~5 min later, the wake lands in a session the client never reads:
22:35:15 INFO aiohttp.access: 127.0.0.1 "POST /v1/chat/completions HTTP/1.1" 200 873 "-" "Python/3.12 aiohttp/3.14.3"
22:35:15 INFO agent.conversation_loop: Turn ended: reason=text_response(finish_reason=stop) … response_len=56 session=api-546cc1dbe338afb4
22:35:15 INFO gateway.wake: wake self-post delivered for session api-546cc1dbe338afb4 (attempt 1)
```

The header-less client has no way to fetch that 56-char follow-up: its next request rebuilds history from its own body, and it holds no id to resume.

**After (head `09e842ef6c`):** forced-sync fallback, result in the same response.

```
gateway log:
22:21:55 INFO tools.delegate_tool: delegate_task: session id api-546cc1dbe338afb4 is bound but not wake-capable (fingerprint-derived for a header-less client — the wake self-post cannot deliver where the client will read it, #98619) — running the batch synchronously instead.
22:21:55 INFO tools.delegate_tool: delegate_task: async delivery unsupported on this session runtime; running the batch synchronously instead.
22:21:55 INFO agent.turn_context: conversation turn: session=20260906_222155_7501ab … platform=subagent … msg='Reply with exactly the single word PONG and nothing else'
22:23:32 INFO agent.tool_executor: tool delegate_task completed (97.81s, 832 chars)

HTTP 200 to the header-less client (after 421 s):
"content": "The delegate_task completed successfully, returning \"PONG\" as requested."
```

No `async_delegation` dispatch, no wake self-post on the head side.

**Tests (same venv, `.[dev,messaging]`):** `pytest tests/tools/test_delegate_apiserver_background.py "tests/gateway/test_api_server.py::TestSessionIdHeader" -q` → **head: 7 passed**; **base with the head's test files copied over: 3 failed / 4 passed**, the three being exactly the new regressions (`test_apiserver_derived_session_id_stays_synchronous`, `test_apiserver_wake_capability_unbound_fails_closed`, `test_wake_capability_follows_session_id_provenance`).

**Where this stands against current `main` (`6b2d4faf`, 2026-09-06) — for the rebase:** the branch is marked conflicting, and the patch coordinates have moved rather than the bug being fixed.

- The wake gate now lives in `tools/delegate_tool_dispatch.py:202` `_resolve_async_wake_sid()`; the `if origin_wake_sid:` at `:221` still dispatches in the background on the bare existence of a bound id. `tools/delegate_tool.py` is a façade now.
- Origin capture is `tools/delegate_tool_dispatch.py:69` `_capture_origin()` (no wake-capability captured).
- The chat-completions route (where `provided_session_id` / `_derive_chat_session_id` live) moved to `gateway/platforms/api_server_openai_routes.py:457` / `:485`.
- `grep -rn wake_capable` on `main` returns nothing — the concept from this PR is absent there, so the same fix shape (thread a `wake_capable` flag from the route into `set_session_vars`, read it fail-closed in `_resolve_async_wake_sid`) still applies, just at the new locations.
- `gateway/wake.py:116` still raises `RuntimeError("wake self-post requires API_SERVER_KEY…")`, so case ① is unchanged on `main` too.

**Not covered:** unauthenticated server (case ①) — unreachable with the current startup guard; the "with `X-Hermes-Session-Id`" control group in the live run (only exercised by the PR's own tests here); macOS/Windows; providers other than a local Ollama. Timings above are CPU inference and irrelevant to the fix.

Happy to re-run the same trace against a rebased head.

---
*Verified by Sho (human) + Alpha (AI agent) on a real, isolated hermes-agent `api_server` setup.*

andrexibiza (commented):
## Review — fix(gateway): require wake-capable session provenance for background delegate_task

**Verdict: REQUEST_CHANGES** — the diagnosis and the fix direction are right, the diff itself is clean and well-tested, but the PR is rebase-stale against current `main`: the code it edits has been refactored into different files, and the underlying bug from #98619 is still live in the post-refactor code on `main` today. This needs to be re-implemented against the current tree (or re-based), not merged as-is.

### What is solid (verified at the head SHA `09e842ef6c`, base `dce2ecb8a`)

- **The reported loss is real, and the causal chain holds:**
  - Header-less chat-completions requests derive the session id from the message fingerprint (`api_server.py:_derive_chat_session_id`, base:5164) — no explicit `X-Hermes-Session-Id`.
  - The wake self-post hard-requires `API_SERVER_KEY`: `gateway/wake.py:116` raises `RuntimeError("wake self-post requires API_SERVER_KEY…")` when unset, and session continuation via the header is 403-gated on that same key (base:5116-5129). So on an unauthenticated server the wake fails after the subagents already ran — result stranded server-side.
  - `wake_capable=("1" if provided_session_id else "")` correctly keys off the trimmed header value (`provided_session_id = request.headers.get(...).strip()`, base:5115), so a whitespace-only header can't slip through as wake-capable.
- **The gate shape is right:** `if _wake_sid and _origin_wake_capable:` → background; `elif _wake_sid:` → honest log and fall through to the existing forced-sync note. The flag is captured pre-child-construction (same clobber-proof timing as `_origin_wake_sid`), and the `except Exception → False` path fails closed.
- **`clear_session_vars` covers the new var** through the shared `_VAR_MAP` list, so the api-server `finally` reset can't leave a stale wake-capable flag.
- **Tests are targeted and meaningful**: explicit-id → `"1"`; header-less → `""`; unbound/undeclared → synchronous with the honest note. They model the real clobber (child build rewrites `HERMES_SESSION_ID`) via the fixture.

### Blocking: rebase-stale against current `main`, where the bug is still live in different code

**1. The PR conflicts with current `main` in 8 regions across 4 files.** Current `main` tip is `693641aa8b`. Since this PR's base (`dce2ecb8a`, 2026-08-30) the tree has moved:
- `gateway/platforms/api_server.py` was heavily refactored (5270+ → 3965 LOC); the chat-completions handler now lives in `gateway/platforms/api_server_openai_routes.py` (`_handle_chat_completions`), and the two `wake_capable=(...)` call sites this PR edits no longer exist in that form.
- The delegate dispatch moved out of `tools/delegate_tool.py` (now 708 LOC) into `tools/delegate_tool_dispatch.py`; the wake gate this PR patches is now `_resolve_async_wake_sid` there.
- `gateway/session_context.py` was rewritten (400+ → 193 LOC); `set_session_vars`/`_VAR_MAP`/`clear_session_vars` are reshaped.

A merge of this branch would need to re-apply the semantics onto all of that. `git merge-tree` against the current base reports conflicts in both production files plus the two test files.

**2. The bug the PR fixes is still present on current `main`.** `main`'s `_resolve_async_wake_sid` (`tools/delegate_tool_dispatch.py:202`) still returns `origin_wake_sid` (background dispatch) whenever `async_delivery_supported()` is false and *any* session id is bound — with no check on whether that id is a client-addressable explicit id or a fingerprint-derived one:

```python
if origin_wake_sid:
    logger.info(..., origin_wake_sid)
    return origin_wake_sid
return None
```

And the fingerprint-derived id still flows into `origin_wake_sid` on main: the header-less path still derives `session_id` (`api_server_openai_routes.py`, `_handle_chat_completions`), `_bind_api_server_session` binds `chat_id=session_id` (main:3647), and `_current_origin_session_id` (`tools/async_delegation.py:512`) returns that bound `HERMES_SESSION_CHAT_ID` when `platform == "api_server"`. So a header-less client on an unauthenticated server still gets background dispatch with a wake that hard-fails — the exact #98619 loss, on current main.

So this is not a superseded-fix situation: the PR is the right fix, aimed at the pre-refactor code that main no longer has.

### Suggested path

Re-implement (or carefully re-base) against current `main`:

- Port the provenance flag into the post-refactor homes: `_handle_chat_completions` in `api_server_openai_routes.py` decides `"1"` vs `""` from `provided_session_id`; thread it through `run_kwargs` / `_spawn_stream_agent` / `_run_agent` / `_bind_api_server_session` in the current `api_server.py`; add the `wake_capable` ContextVar + `_VAR_MAP` entry + `clear_session_vars` reset to the current `session_context.py`.
- Port the gate into `tools/delegate_tool_dispatch.py`: `_capture_origin` should capture the capability flag next to `origin_wake_sid`, and `_resolve_async_wake_sid` should require both before returning a wake target (return `None` → sync-with-note otherwise).
- Re-run the two test files against the current head — they still exist on `main` (`tests/gateway/test_api_server.py`, `tests/tools/test_delegate_apiserver_background.py`) — adapting the new cases to the current architecture (e.g. `_spawn_stream_agent`/`run_kwargs` threading instead of direct `_run_agent` kwargs).

Nothing in this diff is wrong on its own — it is a correct fix against a baseline that has since moved out from under it.

andrexibiza:
Correction to review #5127266065:

The landing disposition remains **REQUEST_CHANGES**, but two statements in my review need to be narrowed.

1. **The no-key failure is not reachable through normal current API-server startup.** Current `main` requires a non-empty, strong `API_SERVER_KEY` in `gateway/platforms/api_server.py:_api_key_passes_startup_guard()` and refuses to start otherwise (`:3786-3815`, called by `connect()` at `:3824`). `gateway/wake.py` still raises when its adapter has no key, but that is a defensive/manual-wiring path, not a normal live deployment path. The live current-main defect is the authenticated header-less case: a fingerprint-derived id is accepted as a wake target, the self-post succeeds, and the plain OpenAI-compatible client does not resume/read that session.

2. **The PR is not clean on provenance authority.** At head `09e842ef6c`, `gateway/session_context.py:set_session_vars()` defaults `wake_capable` to `"1"` (`:259`) and unconditionally binds it (`:311`); `gateway/platforms/api_server.py:_bind_api_server_session()` also defaults it to `"1"` (`:7201`). That means a binder which omits provenance receives wake authority by default. The new `test_apiserver_wake_capability_unbound_fails_closed` does not exercise that public omission: it calls `set_session_vars()` without the argument, then directly mutates the private ContextVar to `_UNSET` (`tests/tools/test_delegate_apiserver_background.py:176-187`). It proves the consumer handles a fabricated `_UNSET` state, not that an omitted declaration fails closed.

The reimplementation on current `main` must therefore:

- preserve the authenticated header-less fix in the post-refactor owners (`gateway/platforms/api_server_openai_routes.py`, `gateway/platforms/api_server.py`, `gateway/session_context.py`, and `tools/delegate_tool_dispatch.py`);
- make the central capability binding default-deny (`""` or `_UNSET`), with explicit authorization only from audited producers that have a client-resumable return address; and
- replace the private-ContextVar test with a public omission regression, while retaining positive controls for explicit header-backed sessions and any genuinely resumable API entry points.

The prior review’s broader conclusion still stands: current `main` is conflicting with this branch, and the provenance fix must be rebuilt against the post-refactor code. No merge should proceed from this head.

liuhao1024:
Rebuilt against current `main` (`693641aa8b`) at head `33308c32fd` — thank you both for the deep reviews and the correction, and @shojikumaru for the real-setup verification and the post-refactor coordinate map, which made this rebuild straightforward. Addressing review #5127266065 and its correction point by point:

**1. Fix ported into the post-refactor owners.** The provenance flag is decided in `_handle_chat_completions` (`gateway/platforms/api_server_openai_routes.py`) from `provided_session_id`, threaded through `run_kwargs` → `_spawn_stream_agent`/`_run_agent` (`gateway/platforms/api_server.py`) into the single bind chokepoint `_bind_api_server_session()`; `wake_capable_session()` lives in the reshaped `gateway/session_context.py`; and the gate is in `tools/delegate_tool_dispatch.py` — `_capture_origin()` captures the capability next to `origin_wake_sid` (pre-child-construction, read error ⇒ not capable), `_resolve_async_wake_sid()` now requires **both** the bound id and the declared capability, with an honest "bound but not wake-capable → synchronous" log line.

**2. Central binding now default-deny.** `set_session_vars(wake_capable=...)` and `_bind_api_server_session(wake_capable=...)` both default to *undeclared* — an omitted declaration binds `_UNSET` and `wake_capable_session()` returns True only for the literal `"1"`. The var is deliberately outside `_VAR_MAP`: no `os.environ` fallback (a leaked `HERMES_SESSION_WAKE_CAPABLE=1` grants nothing — unit-tested) and no subprocess-env-bridge export. Explicit `"1"` is passed only by audited producers whose client can address the id again: explicit `X-Hermes-Session-Id` (403-gated on `API_SERVER_KEY`), native `/api/sessions/{session_id}/chat[/stream]` (id in the request path), and `/v1/runs` (body id / response-chain id / declared key / `run_id` fallback — all client-held, and a wake-injected turn lands in session history a later run on the same id loads). `/v1/responses` fresh-uuid sessions are **not** opted in: I could not audit that surface as a wake-resumable return address (its history comes from the response store, not the session the self-post wakes), so it now honestly falls back to sync instead of dispatching into an unaudited channel.

**3. Private-ContextVar test replaced with a public omission regression.** `test_apiserver_wake_capability_omission_fails_closed` binds via the public `set_session_vars()` *without* the flag and proves the batch stays synchronous — no surgery on private state. Positive controls: the explicit-header route test (`TestSessionIdHeader::test_wake_capability_follows_session_id_provenance`, "1" vs "" by header provenance), the declared-"1" dispatch control in the delegate tests, and the messaging-gateway surfaces are untouched (they take the `async_delivery_supported()` path, which never reaches the wake gate). Reader tri-state + env-no-fallback is covered by `test_wake_capable_session_reader_is_default_deny`.

**Godfile note (from the first review):** the dispatch extraction upstream already performed resolves P1-2 for the delegate side — the gate now lands in `tools/delegate_tool_dispatch.py` (406 LOC, the bounded owner) while `delegate_tool.py` stays a 708-LOC façade, and the `api_server.py` delta is confined to the chokepoint signature/throughput rather than new policy hunks.

**Credit:** the commit carries `Reported-and-investigated-by: shojikumaru (Sho + Alpha) via #98619` and the PR body credits the first-hand report and the real-setup verification explicitly.

Relevant tests: `tests/tools/test_delegate_apiserver_background.py` (5), `TestSessionIdHeader` (3), plus a 292-test regression sweep across `test_api_server*.py`, `test_session_context_inheritance.py`, `test_wake_delivery.py`, and the delegate/async-delegation suites — all green.

andrexibiza:
## Re-review — one remaining blocking contract gap

**Verdict: REQUEST_CHANGES.** The rebuilt head `33308c32fdc9e78004308d25f8dc2af7481767d2` is materially improved: the central provenance primitive is now default-deny, the explicit OpenAI-header and native session producers are appropriately gated, and `/v1/responses` remains conservatively synchronous. One blocking defect remains in the `/v1/runs` opt-in.

### P1 — `/v1/runs` grants detached-delivery authority to continuations that cannot consume the completion

`gateway/platforms/api_server_runs.py:486-494` unconditionally binds every `/v1/runs` execution with `wake_capable="1"`. That includes `previous_response_id` continuations, declared `X-Hermes-Session-Key` resolution, explicit `session_id` requests carrying caller-supplied `conversation_history`, and the generated `run_id` fallback.

The completion path does not update the Runs API. For an API-server `async_delegation`, `gateway/run_notifications.py:842-864` deliberately calls `persist_delegation_delivery`, not a wake/self-post. `gateway/wake.py:74-100` appends a SessionDB row with `display_kind="async_delegation_complete"`; it does not update `_run_statuses`, the run SSE queue, or the ResponseStore.

The parent run can settle before the child. `gateway/platforms/api_server_runs.py:558-605` records only the parent run’s immediate result, emits the terminal SSE sentinel, and retires the live transport. `GET /v1/runs/{run_id}` reads that parent status (`:659-663`), while `/events` drains the parent queue and closes on the sentinel (`:666-705`). A late `subagent.complete` callback is not a durable, reconnectable delegation-result record on those surfaces.

The `previous_response_id` branch makes the mismatch directly observable. Its history comes from the ResponseStore (`gateway/platforms/api_server_runs.py:270-285`), and the SessionDB fallback is skipped when `previous_response_id` is present (`:419-420`). The async completion is written only to SessionDB, so the subsequent response-chain run reloads a stale ResponseStore snapshot and omits the completion. The same class applies when the caller supplies non-empty conversation history or when a declared session key is resolved after the history-load decision.

There is also a transcript-ordering risk for a concurrent continuation: `persist_delegation_delivery()` appends a `role=user` delivery row without a turn-lease holder or active-lease rejection. If a later run has already loaded the session and then writes its own user turn, the delivery can create adjacent user messages and arrive in the wrong order. The existing idle-session delivery test does not exercise this reachable overlap.

This conflicts with `website/docs/user-guide/features/api-server.md:472-486`, which presents the Runs SSE stream as carrying `subagent.complete` outcomes. That may hold while the parent stream is still open; it does not provide a delivery contract for a detached completion that arrives after the parent has closed.

**Required fix:** either keep `/v1/runs` default-denied and retain the synchronous fallback, or make detached completion a real Runs API contract: preserve the originating `run_id` through completion, project completion/failure into durable run status and reconnectable event/result replay, merge the completion into every supported continuation source or deny wake capability for those branches, and add end-to-end regressions for late child completion, reconnect/poll, response-chain continuation, declared-key continuation, and concurrent transcript writes.

### Verified on this head

- Central wake capability is default-deny; omission does not grant authority.
- Explicit OpenAI session-header and native path-addressed session producers are appropriately opt-in.
- `/v1/responses` remains default-denied because its continuation is ResponseStore-based rather than SessionDB-based.
- Local targeted verification: `14 passed` in the provenance/bind suites, followed by `146 passed` across gateway/delegate/session suites.
- Exact-head GitHub CI is green; the PR is mergeable.

This is narrower than the earlier review: the provenance gate is in the right shape, but `/v1/runs` still grants authority beyond the delivery contract it actually implements.

andrexibiza:
## Precision update to the re-review

The default-deny finding remains resolved. The remaining blocker is narrower than “every `/v1/runs` completion is unobservable”: an explicit `session_id` run with no caller-supplied history can expose the persisted delivery row through `GET /api/sessions/{id}/messages`, and the run status does retain the selected `session_id`.

The blocking defect is that the PR grants `wake_capable="1"` **unconditionally**, including continuation paths whose actual history source cannot consume that SessionDB row:

- `previous_response_id` continuations load the ResponseStore snapshot (`gateway/platforms/api_server_runs.py:270-285`) and skip the SessionDB fallback when `previous_response_id` is present (`:419-420`). The async completion is persisted only to SessionDB (`gateway/run_notifications.py:842-864`; `gateway/wake.py:74-100`), not merged into the ResponseStore chain. The next response-chain run therefore resumes from stale history.
- A declared session key is resolved after the history-load decision (`:423-427`), so a header-only continuation can select the same session without loading the persisted delivery row.
- Caller-supplied non-empty `conversation_history` also bypasses the SessionDB load, leaving the detached completion outside the history the next run actually uses.
- Independently of continuation history, a child that completes after the parent run closes is not projected into the originating run's status or reconnectable `/v1/runs/{run_id}/events` stream. The parent terminal status and SSE sentinel are emitted at `:558-605`; the completion watcher writes a separate SessionDB delivery row instead.

So the required change is not necessarily to remove wake capability from every `/v1/runs` request. It is to make the opt-in conditional on a continuation path with a proven completion consumer, or to project the completion into the run/ResponseStore surfaces. At minimum, default-deny `previous_response_id`, header-only declared-key, and caller-history branches until they have explicit delivery coverage. Add regressions that dispatch a background delegate, let the parent settle, complete the child, and then verify the result through each supported continuation and reconnect path.

The independent verification still agrees on the broader shape: the ContextVar capability is default-deny, explicit OpenAI-header and native path-addressed sessions are appropriately gated, `/v1/responses` remains denied, and the exact head/CI are healthy.

andrexibiza:
## Adjudication of the independent `/v1/runs` pass

The default-deny implementation is not in dispute. The remaining blocker is narrower than “no `/v1/runs` client can ever read a completion”: a client that already knows the selected raw session ID can read the persisted SessionDB delivery row through `GET /api/sessions/{id}/messages`.

That does not make the unconditional opt-in safe for every supported `/v1/runs` continuation:

- `previous_response_id` continuations reconstruct history from ResponseStore (`gateway/platforms/api_server_runs.py:270-285`) and deliberately skip the SessionDB fallback when `previous_response_id` is present (`:419-420`). The completion is written only to SessionDB (`gateway/run_notifications.py:842-864`, `gateway/wake.py:74-100`), so the next response-chain run does not consume it.
- Header-only declared-session-key selection is resolved after the history-load decision (`:423-427`), so selecting the same session does not itself prove that the persisted delivery row enters the next run’s context.
- Caller-supplied non-empty `conversation_history` likewise bypasses the SessionDB load.
- A late completion is not projected into the originating run’s status or reconnectable `/v1/runs/{run_id}/events` stream; the parent run closes its transport before the detached completion is persisted.

The independent pass therefore validates the explicit-session/session-messages path, but does not clear the continuation and Runs-API contract gap. The required change remains: make `wake_capable="1"` conditional on a continuation path with a proven completion consumer, or project/merge the completion into each relevant run/ResponseStore surface, with regressions for these branches.

andrexibiza:
## Final re-review adjudication

The independent lanes agree on the important split:

- **PASS:** the central wake-capability primitive is genuinely default-deny. Omission, empty values, and environment injection do not authorize a wake; only literal `"1"` does. Explicit `X-Hermes-Session-Id`, native path-addressed sessions, and `/v1/responses` behavior are correctly classified.
- **Still blocking:** the `/v1/runs` producer unconditionally supplies `wake_capable="1"` at `gateway/platforms/api_server_runs.py:484-496`.

The remaining issue is not that no client can ever read a completion. A client that already knows the selected raw session ID can read the persisted SessionDB delivery row through `GET /api/sessions/{id}/messages`. The issue is that the unconditional opt-in covers `/v1/runs` continuations whose actual consumer is different:

- `previous_response_id` reloads ResponseStore history (`:270-285`) and skips the SessionDB fallback (`:419-420`); the async completion is persisted only to SessionDB (`gateway/run_notifications.py:842-864`, `gateway/wake.py:74-100`).
- Header-only declared-session-key selection resolves after the history-load decision (`:423-427`), so the selected session does not itself prove that the delivery row enters the next run’s context.
- Caller-supplied non-empty `conversation_history` similarly bypasses SessionDB loading.
- A child completing after the parent closes is not projected into the originating run status or reconnectable `/v1/runs/{run_id}/events`; those surfaces describe and close the parent run (`:547-605`, `:659-705`).

**Required change:** make `wake_capable="1"` conditional on a continuation path with a proven completion consumer, or project/merge the completion into the relevant Run/ResponseStore surfaces. Add regressions for `previous_response_id`, header-only declared-key, caller-supplied history, and late completion after parent settlement.

andrexibiza:
## Re-review precision update — two concrete `/v1/runs` blockers

The latest exact-head HTTP probes sharpen the remaining finding. The generated `run_id` fallback is resumable when a client posts that returned ID back as `body.session_id`; that branch loads SessionDB history. The unconditional opt-in is still invalid for two other supported `/v1/runs` paths:

### P1 — declared `X-Hermes-Session-Key` does not load the selected session history

At `gateway/platforms/api_server_runs.py:419-427`, the handler attempts `_conversation_history_for_session(session_id)` before resolving a declared session key. A request supplying only `X-Hermes-Session-Key` therefore reaches the load gate with `session_id=None`, resolves the declared session afterward, and runs without its persisted SessionDB delivery history. The PR then grants that route `wake_capable="1"` at `:485-494`.

An exact-head HTTP probe resolved the key to `declared-session`; the request returned `202`, but `_conversation_history_for_session` was never awaited. A detached completion persisted by `gateway/wake.py:74-100` is therefore not available to the next same-key run.

**Required fix:** resolve the body/response-chain/declared-key/run-id session identity before loading history, then load the resolved declared session. Add an HTTP regression that persists an async-delegation delivery row and verifies the next same-key run receives it.

### P1 — `previous_response_id` continuations are pinned to a stale ResponseStore snapshot

`gateway/platforms/api_server_runs.py:248-285` resolves `previous_response_id` from the ResponseStore. The later history-load condition at `:419-420` explicitly skips SessionDB whenever `previous_response_id` is present, even when the stored session ID is known. Async completion is persisted only to SessionDB (`gateway/run_notifications.py:842-864`; `gateway/wake.py:74-100`), not merged into the ResponseStore chain.

An exact-head HTTP probe seeded a response-chain snapshot and a newer SessionDB delivery row; the next `previous_response_id` request returned `202`, but the history loader was never awaited and the agent received only the stale ResponseStore transcript.

**Required fix:** merge/reload current SessionDB history for the stored response-chain session, or deny wake capability for this branch until that delivery contract exists. Add a regression covering completion persistence between the prior response and the next `previous_response_id` request.

The broader result remains unchanged: the central capability is genuinely default-deny, explicit header/native paths are properly classified, `/v1/responses` remains denied, and the returned `run_id` fallback itself is resumable. The remaining defect is the unconditional `/v1/runs` opt-in across these two unsupported continuation paths.

andrexibiza:
## Additional blocking finding — compression-rotated `/v1/runs` sessions strand completions

The default-deny provenance gate remains sound. A new exact-head reproduction adds another `/v1/runs` delivery blocker:

`gateway/platforms/api_server_runs.py:421-432,486-494` binds the original session ID as the wake target. If the parent run compresses/rotates before its detached delegate finishes, `tools/delegate_tool_dispatch.py:371-383` preserves that original ID as `origin_session_id`. The API-server completion path then calls `persist_delegation_delivery()` with that stale ID (`gateway/run_notifications.py:842-864`; `gateway/wake.py:74-100`).

`persist_delegation_delivery()` appends directly with `db.append_message(session_id, ...)`; it does not resolve/adopt the live compression continuation. The SessionDB write guard rejects the closed parent with `CompressionSessionClosedError` (`hermes_state_messages.py:284-291`). The watcher retries, but no completion is written to the live tip. Normal message retrieval resolves the original ID to the continuation tip (`gateway/platforms/api_server.py:2919`; `hermes_state_messages.py:691-705`), so the client cannot recover the stranded row through `/api/sessions/{id}/messages`, and the run status/SSE surfaces do not carry it either.

**Required fix:** resolve/adopt the live compression continuation before appending the delivery row, update the client-visible run/session reference to that target, and add an end-to-end `/v1/runs` regression where the parent compresses before the detached child completes and the result is retrieved afterward.

This is in addition to the declared-session-key and `previous_response_id` continuation blockers already recorded.

liuhao1024:
All three blockers are addressed at head `4be09076eb` (two commits on top of `33308c32fd`).

**P1 — declared `X-Hermes-Session-Key` does not load the selected session history.** `_handle_runs` now resolves the full session identity (body `session_id` > response chain > declared key > `run_id` fallback) *before* the history-load decision, then loads SessionDB history for whichever session was actually selected — a header-only declared-key run now loads its declared conversation's history, so a persisted async-delegation delivery row reaches the next same-key run's context. Regressions: `test_declared_session_key_loads_selected_session_history` seeds a delivery row and verifies the next header-only run consumes it (`get_messages_as_conversation("declared-session")` → the run's `conversation_history`); `test_declared_key_without_live_session_row_loads_nothing` keeps the no-live-row case fresh (no load, `run_id` fallback).

**P1 — `previous_response_id` continuations pinned to a stale ResponseStore snapshot.** Took the deny option from your two: a `previous_response_id` continuation consumes its ResponseStore snapshot and can never see a SessionDB delivery row, so it now binds `wake_capable=""` and `delegate_task` keeps its synchronous fallback on that branch until a merge contract exists for the chain — the same default-deny the review already accepted for `/v1/responses` (ResponseStore-based continuation). The snapshot itself remains the run's history, unchanged. Regressions: `test_previous_response_id_continuation_is_not_wake_capable` seeds a response-chain snapshot plus a stored session id and asserts the wake binding is denied while the snapshot history is still used; `test_plain_run_grants_wake_capability` is the control (a body session id the client can address again stays `"1"`).

**Compression-rotated `/v1/runs` sessions stranding completions.** `persist_delegation_delivery` now adopts the live continuation tip before appending — the same canonical `resolve_resume_session_id` resolution `GET /api/sessions/{id}/messages` reads use — so the delivery row lands on the tip the read surfaces resolve, instead of being rejected on the closed parent forever (the watcher retry loop no longer re-writes the stale id). `/v1/runs` adopts that tip for a client-addressed session id (body id or declared key) too: the history load, the turn, and the wake binding all target the live session rather than writing into the closed parent. End-to-end regression `test_compressed_parent_delivery_row_lands_on_live_tip`: the parent compresses (`end_session(..., "compression")` + live child), the detached delivery persists against the stale origin id, and the next same-id run loads the delivery row from the tip and binds the tip.

Local: 272 passed across `test_api_server_runs.py`, `test_wake_delivery.py`, `test_delegate_apiserver_background.py`, `test_api_server.py`, `test_api_server_declared_conversation.py`, and the background-notification / kanban wake suites; ruff clean.

andrexibiza:
## New blocker after the latest rebuild — caller-supplied history still receives wake authority

The rebuild correctly closes the declared-key ordering defect, denies `previous_response_id`, and adopts compression tips. One provenance hole remains at exact head `4be09076ebecaff852a4344b797e56081f1c3949`:

`gateway/platforms/api_server_runs.py:453-460` loads SessionDB history only when `not conversation_history`. But `_RunLaunch.wake_capable` is assigned as `not previous_response_id` at `:477-480`, and `_run_agent_sync` grants `wake_capable="1"` whenever that boolean is true at `:519-530`.

Therefore a request with an explicit `session_id` plus non-empty caller-supplied `conversation_history` skips the SessionDB load yet is still marked wake-capable. A detached API-server completion is persisted only as a SessionDB delivery row (`gateway/run_notifications.py:842-864`; `gateway/wake.py:74-100`). The subsequent run that supplies the same snapshot as `conversation_history` does not consume that row, so the opt-in has promised a completion consumer that this path does not use.

This was reproduced against the exact head: the run reached the agent with `wake_capable="1"` while `_conversation_history_for_session` remained uncalled. The current added tests cover explicit session IDs with empty history, declared-key loading, `previous_response_id` denial, and compression-tip persistence, but do not cover this non-empty-history branch.

**Required fix:** either deny wake capability whenever caller-supplied `conversation_history` is authoritative, or merge/load the current SessionDB transcript before granting the capability. Add an HTTP regression with `session_id` + non-empty `conversation_history` that proves the binder receives `wake_capable=""` (or proves the persisted completion is merged into the actual next-turn history).

### What now holds

- Default-deny remains proof-carrying.
- Explicit OpenAI-header and native path-addressed routes remain properly gated.
- Declared-key history ordering is repaired.
- `previous_response_id` is now denied as it should be.
- Compression-tip delivery is now adopted and tested.
- The canonical affected suites are green: **233 passed**.

liuhao1024:
Confirmed — good catch. `wake_capable` was derived from `not previous_response_id` alone, so an explicit `session_id` + non-empty caller-supplied `conversation_history` skipped the SessionDB load yet still granted wake authority, promising a completion consumer that path never provides.

Fixed in dc1e6657 by taking the deny branch you listed first: the capability is now computed as `not previous_response_id and not conversation_history`, pinned **before** `_conversation_history_for_session` can overwrite the variable (the load happens later on the same name, so judging after it would no longer distinguish caller-supplied from loaded history). Caller-supplied history is authoritative for the turn and never reads the SessionDB delivery row a detached completion persists, so it is denied on the same default-deny contract as `previous_response_id` continuations — merging the persisted transcript into a caller-owned snapshot would need a merge contract this chain deliberately does not have yet.

Added the requested HTTP regression (`test_caller_supplied_history_with_session_id_is_not_wake_capable`): `session_id` + non-empty `conversation_history` proves the binder receives `wake_capable=""`, `_conversation_history_for_session` stays uncalled, and the caller snapshot reaches the agent untouched. Control case (`test_plain_run_grants_wake_capability`) still asserts `wake_capable="1"`. Affected suites green: `tests/gateway/test_api_server_runs.py` + `tests/tools/test_delegate_apiserver_background.py` (69 passed), `tests/gateway/test_api_server.py` (111 passed).

gateway/platforms/api_server_openai_routes.py
```diff
@@ -494,7 +494,13 @@ async def _handle_chat_completions(self, request: "web.Request") -> "web.Respons
...
             user_message=user_message, conversation_history=history,
             ephemeral_system_prompt=system_prompt, session_id=session_id,
-            gateway_session_key=gateway_session_key, **agent_overrides, route=route)
+            gateway_session_key=gateway_session_key, **agent_overrides, route=route,
+            # #98619: only an explicitly provided X-Hermes-Session-Id is wake-capable (the
+            # header is 403-gated on API_SERVER_KEY, so the wake self-post can authenticate
+            # and the client can resume the session by sending it again). A fingerprint-derived
+            # id from a header-less client is NOT: delegate_task keeps its forced-sync fallback
+            # there — the wake would hard-fail or land in history that client never reloads.
+            wake_capable=("1" if provided_session_id else ""))
```
andrexibiza, line 510: **[P1] Resolve the explicit header to the same live session as the delivery writer before granting resumability.**

This grants `wake_capable="1"` to a pre-compression parent ID, but the preceding header branch still uses that ID directly for both `get_messages_as_conversation()` and `_run_agent`. The new `persist_delegation_delivery()` instead resolves the captured parent to its child before writing. For a streamed turn that rotates `P → C`, response headers have already exposed `P` and the final Chat Completions chunk does not expose `C`. A subsequent request repeating `P` therefore reads the parent and misses the delivery on `C`, despite this opt-in. The getter has no implicit tip resolution; the turn-lease reload is conditional on waiting, so idle continuation is not repaired there.

The caller-history correction is separate and resolved. This is the remaining writer/reader interlock, not a regression attributed to that latest commit. [#13437](https://github.com/NousResearch/hermes-agent/pull/13437) is the existing compressed-header routing carrier: consolidate/land its current-main semantics with preserved provenance rather than adding a competing lineage policy. Resolve the effective session before history loading and binding, while preserving the external ID contract.

Add a real-SessionDB HTTP regression: close `P` by compression, create child `C`, persist a late delegation through origin `P`, and continue `/v1/chat/completions` with header `P`. Assert `_run_agent(session_id="C", conversation_history=...)` includes the completion; exercise both streaming and JSON plus the non-rotated control. Keep async delegation as a delivery row, not an unsolicited self-posted agent turn.

andrexibiza (commented):
## Re-review at `dc1e6657eb2ad638ef13159479e7beffb16613df`

**The caller-supplied-history finding in [the preceding review comment](https://github.com/NousResearch/hermes-agent/pull/98630#issuecomment-5564754965) is resolved.** The change described in [5564853027](https://github.com/NousResearch/hermes-agent/pull/98630#issuecomment-5564853027) computes the capability before the SessionDB load can replace `conversation_history`. Non-empty caller history and `previous_response_id` are denied; a plain session-history continuation remains eligible. The new HTTP regression also checks that the caller snapshot is preserved and the DB loader is not called. I am not reopening that finding.

The central default-deny binding, environment isolation, pre-child capability capture, declared-key history-load ordering, and `/v1/runs` compression-tip selection are present in this head. The remaining functional concern is a different producer of the same capability, described in the inline comment.

### P1 — Complete the compression delivery/read interlock for explicit-header Chat Completions

The new delivery writer resolves a captured parent ID to its live child, but the explicitly opted-in Chat Completions reader still loads and binds the supplied parent directly. A streamed turn can rotate `P → C` after its response headers have advertised `P`; the success SSE terminator does not publish the rotated ID. A late delegation delivery now lands on `C`, while the next `/v1/chat/completions` request with `X-Hermes-Session-Id: P` reads `P` and omits that result, yet still receives `wake_capable="1"`.

Exact-head evidence:

- [Producer, direct history read, and capability grant](https://github.com/NousResearch/hermes-agent/blob/dc1e6657eb2ad638ef13159479e7beffb16613df/gateway/platforms/api_server_openai_routes.py#L452-L503).
- [Delivery writer now follows the continuation before appending](https://github.com/NousResearch/hermes-agent/blob/dc1e6657eb2ad638ef13159479e7beffb16613df/gateway/wake.py#L74-L117).
- [The SessionDB getter selects the requested session; it does not implicitly resolve the tip](https://github.com/NousResearch/hermes-agent/blob/dc1e6657eb2ad638ef13159479e7beffb16613df/hermes_state_messages.py#L734-L754).
- [SSE headers precede completion, and the finish frame does not carry `result.session_id`](https://github.com/NousResearch/hermes-agent/blob/dc1e6657eb2ad638ef13159479e7beffb16613df/gateway/platforms/api_server_openai_routes.py#L599-L677). The turn-lease fallback [reloads the tip only after waiting](https://github.com/NousResearch/hermes-agent/blob/dc1e6657eb2ad638ef13159479e7beffb16613df/agent/turn_facade_lease.py#L277-L305), so it is not an unconditional repair for an idle resumed session.

This is **not a claim that the latest caller-history commit introduced the underlying compressed-header bug**, nor that the delivery is absent from every read surface. Native message polling can resolve the child. It is an incomplete dependency of this PR's explicit-header resumability guarantee: the client continuing through its original Chat Completions route does not consume the newly redirected delivery.

**Existing carrier:** [#13437](https://github.com/NousResearch/hermes-agent/pull/13437) already implements compressed-header history/read-write routing and stable external identity. I inspected its patch at `5b22937e615e12123c50e54258a59252a39c721c`; it remains open, unmerged and conflicting, and targets the old `api_server.py` owner. Use that existing repair's semantics/tests and contributor provenance in the post-refactor route, or make its current-main landing an explicit dependency; do not create another competing resolver. The delivery producer and continuation consumer must select the same live session, while preserving the client-visible identity contract and the no-self-post delivery behavior.

**Required regression:** use a real SessionDB with closed compression parent `P` and live child `C`; persist a late delegation through origin `P`; then send an authenticated Chat Completions continuation with header `P`. Assert that the actual `_run_agent` arguments select `C` and include the delivery row. Cover streaming as well as JSON, including a first streamed turn that rotates after headers are prepared. Keep the existing non-rotated positive control and header-less denial.

### Verification and surviving-commit CI

The PR is open, non-draft and mergeable at the reviewed head. Compared with `main`/merge-base `693641aa8b4359c602283bdbbc14041e03bc47bc`, it is four commits ahead and zero behind, with nine changed files.

I executed an isolated source-decision witness against the retrieved head's delivery/read and Runs capability logic: a real minimal SQLite stand-in, with adapter helpers stubbed, **not the complete Hermes SessionDB/runtime**. It reproduced `delivery→C`, then `header P→session P / wake=1 / result absent`; the `header C` control consumed the result and the `/v1/runs` resolver selected `C`. Eight caller-history/response-chain/session-selection combinations matched the new deny rule. These are bounded local checks, not a claim to have rerun repository pytest or a live-provider end-to-end test. The contributor's reported suite counts are not independent execution receipts from this review.

| Surviving commit | CI | Docker | Nix |
|---|---|---|---|
| `33308c32fdc9` | [success](https://github.com/NousResearch/hermes-agent/actions/runs/34074670021) | [success](https://github.com/NousResearch/hermes-agent/actions/runs/34074669195) | [success](https://github.com/NousResearch/hermes-agent/actions/runs/34074669287) |
| `b9736f12f8fe` | **[cancelled](https://github.com/NousResearch/hermes-agent/actions/runs/34079354898)** | [success](https://github.com/NousResearch/hermes-agent/actions/runs/34079354445) | [success](https://github.com/NousResearch/hermes-agent/actions/runs/34079354370) |
| `4be09076ebec` | [success](https://github.com/NousResearch/hermes-agent/actions/runs/34079677847) | [success](https://github.com/NousResearch/hermes-agent/actions/runs/34079677418) | [success](https://github.com/NousResearch/hermes-agent/actions/runs/34079677416) |
| `dc1e6657eb2a` | [success](https://github.com/NousResearch/hermes-agent/actions/runs/34081738675) | [success](https://github.com/NousResearch/hermes-agent/actions/runs/34081738204) | [success](https://github.com/NousResearch/hermes-agent/actions/runs/34081739992) |

The unfiltered exact-SHA Actions query for `b9736f12f8feb2775f55435aceba71a067eb56e4` returns these three runs and no alternate successful CI run. Cancellation is **not a demonstrated code failure**, but the current green head does not supply the missing green receipt for that surviving commit. The every-commit-green gate is therefore still unmet.

**Disposition:** the linked caller-history correction is accepted as resolved; the explicit-header compression interlock and the cancelled surviving-commit CI receipt still prevent a complete landing sign-off for [#98619](https://github.com/NousResearch/hermes-agent/issues/98619).

shojikumaru:
Follow-up on the rebuilt branch, as promised.

**Tests, two-sided, on the rebuilt head `4be09076eb` vs. its base (current `main` `693641aa8b`)** — same isolated Linux venv as before (`.[dev,messaging]`, Python 3.12.3):

- head: `pytest tests/tools/test_delegate_apiserver_background.py "tests/gateway/test_api_server.py::TestSessionIdHeader" -q` → **8 passed**
- base with the head's two test files copied over → **5 failed, 5 errors, 3 passed**. The errors are `gateway.session_context` lacking `_SESSION_WAKE_CAPABLE` / `wake_capable_session` (the new default-deny primitive does not exist on `main`), and the failures include `test_apiserver_derived_session_id_stays_synchronous` and `test_wake_capability_omission_fails_closed` — i.e. the header-less path still dispatches in the background on `main`, which is the #98619 loss.

**Live re-run — not completed on the rebuilt head, honestly.** I re-ran the same header-less `curl → /v1/chat/completions → delegate_task(background=true)` trace against `4be09076eb` and against `main`, but with the tool surface on current `main` none of the local Ollama models I have would emit a real `delegate_task` call (qwen3 8B looped and gave up; gemma3-tools 12B wrote the call as fenced text; the Qwen3.6 GGUF has no tool template). So the before/after log evidence in my earlier comment stands for the pre-rebase head `09e842ef6c` only. Note that `dc1e6657eb` (the newest commit) touches only `gateway/platforms/api_server_runs.py` + its test, so the header-less chat-completions production path is the same at `4be09076eb` and `dc1e6657eb`.

If a maintainer wants the live trace on the rebuilt head as well, I can run it with a hosted tool-capable model on request; otherwise the regression tests above cover the same boundary (`TestSessionIdHeader` for header provenance → `wake_capable`, the delegate tests for the dispatch gate).

*Sho (human) + Alpha (AI agent), isolated real `api_server` setup.*

liuhao1024:
Interlock completed at `ef4687ee7a` — the explicit-header Chat Completions continuation now selects the same live session the detached delivery writer writes to, with #13437's stable-identity contract on the wire.

**P1 — compression delivery/read interlock for explicit-header Chat Completions.** `_handle_chat_completions` now resolves the client-provided `X-Hermes-Session-Id` through the same canonical `resolve_resume_session_id` resolution the delivery writer (`gateway/wake.py`) and `/v1/runs` already use, before loading history. The read, the turn binding, and the wake target all adopt the live continuation tip, so a late delegation delivery row persisted on the tip is exactly what the next header continuation consumes — no competing resolver was introduced. Response headers keep echoing the client-provided id on both paths (SSE headers and the JSON response): the external identity stays stable across rotations, and a turn that rotates *after* headers are prepared never changes what the client should re-send — it sends the same id again and the tip resolution finds whatever is live by then. A fingerprint-derived (header-less) turn keeps the existing behavior of reporting the id the agent actually resolved.

**Regressions** (real `SessionDB`, closed compression parent `P` + live child `C`, late delivery row persisted through the stale origin id via `persist_delegation_delivery`):
- `test_provided_session_id_adopts_compression_tip_for_history_bind_and_wake` (JSON): asserts `_run_agent` receives `session_id="child-session"` with the delivery row in `conversation_history` and `wake_capable="1"`, while the response echoes `parent-session` even though the turn reports `session_id: child-session`.
- `test_streamed_session_id_header_adopts_tip_and_echoes_stable_id` (streaming): SSE response headers carry the stable client id prepared before the turn, and the streamed turn still runs against the tip.
- Both fail on the pre-fix head and pass on `ef4687ee7a`; the non-rotated positive control (`test_provided_session_id_loads_history_from_db`, resolver pinned to identity) and the header-less denial (`test_wake_capability_follows_session_id_provenance`) are retained. `236 passed` across `test_api_server.py`, `test_api_server_runs.py`, `test_api_server_declared_conversation.py`, `test_delegate_apiserver_background.py`, and `test_wake_delivery.py`; ruff clean.

**#13437 provenance.** The compression-tip read/write routing and the stable-echo identity contract are manuelschipper's design from #13437; this ports those semantics and test intent onto the post-refactor route owner and credits them in the commit (`Co-authored-by`). Not a competing resolver — the state-layer helper (`resolve_resume_session_id` / `get_compression_tip`) is reused as-is.

**Cancelled CI on `b9736f12f8fe`.** That run was cancelled because the branch moved six minutes later (`4be09076ebec` pushed 03:26 vs `b9736f12f8fe` 03:20) — GitHub's concurrency auto-cancel, not a code failure. The chain is linear (`33308c32fd` → `b9736f12f8` → `4be09076eb` → `dc1e6657eb` → `ef4687ee7a`), so `b9736f12f8`'s changes are fully contained in both successors whose CI/Docker/Nix runs all succeeded; CI for the new head is up.

@shojikumaru — thank you for the two-sided verification; head `8 passed` vs base `5 failed / 5 errors` is exactly the boundary these regressions pin, and your note that `dc1e6657eb` leaves the header-less production path untouched matches the diff (that commit only moved `/v1/runs` capability computation; today's `ef4687ee7a` is the first commit since the rebuild that changes the Chat Completions path). Agreed the regression suites cover the same boundary as the live trace — if a maintainer wants the hosted-model trace on the rebuilt head anyway, that offer stands on your side.

andrexibiza (commented):
## Re-review at `ef4687ee7aaa9fe0a52d96824591bb1027e982e6`

The linked Sho/Alpha two-sided verification is useful evidence for the rebuilt provenance boundary at `4be09076eb`: the targeted head tests pass while the same tests fail/error on `main`, and the report is appropriately explicit that the live provider trace was not rerun on that rebuilt head. Since that comment, however, this branch moved again and `ef4687ee7a` is the first later commit that changes the Chat Completions continuation path, so I reviewed the current exact head rather than treating the `4be09076eb` result as a receipt for it.

### The remaining functional P1 from my `dc1e6657eb2a` review is resolved

The compression delivery/read interlock now composes correctly for explicit-header Chat Completions:

- `gateway/wake.py` resolves a stale captured parent through `resolve_resume_session_id` before persisting the detached delivery row.
- `gateway/platforms/api_server_openai_routes.py` now resolves the client-provided `X-Hermes-Session-Id` through the same live-session resolution **before** loading history, and passes that resolved session to the turn/wake binding.
- JSON and SSE both keep the external identity stable by echoing the client-provided id while the internal read/write/wake target uses the live continuation tip.
- The new real-`SessionDB` regressions cover closed parent `P` → live child `C`, late delivery through stale `P`, continuation read from `C`, and both JSON and streaming response identity. The non-rotated explicit-id control and the header-less default-deny control remain present.

That closes the exact hole I raised at `dc1e6657eb2a`: the detached delivery producer and the explicit-header continuation consumer now select the same live session. The implementation also reuses the existing state-layer lineage resolution rather than introducing a competing resolver, and the `ef4687ee7a` commit carries the #13437/manuelschipper provenance explicitly.

I found no new code blocker in the current diff.

### Exact-head acceptance

Current `main` is still the PR base, `693641aa8b4359c602283bdbbc14041e03bc47bc`; the PR is open, non-draft, mergeable, and zero-behind. Exact current head is green across the three top-level workflows:

- CI: https://github.com/NousResearch/hermes-agent/actions/runs/34084303919 — **success**
- Docker Build, Test, and Publish: https://github.com/NousResearch/hermes-agent/actions/runs/34084303355 — **success**
- Nix flake check: https://github.com/NousResearch/hermes-agent/actions/runs/34084303354 — **success**

One landing gate still remains under the every-surviving-commit-green contract. The linear commit chain is:

`33308c32fdc9` → `b9736f12f8fe` → `4be09076ebec` → `dc1e6657eb2a` → `ef4687ee7aaa`

`b9736f12f8feb2775f55435aceba71a067eb56e4` still has Docker and Nix success but its exact-SHA CI run is **cancelled**:

https://github.com/NousResearch/hermes-agent/actions/runs/34079354898

I accept the explanation that the cancellation was branch-move/concurrency cancellation rather than a demonstrated code failure. What I do **not** accept is treating successor-green as the missing exact-commit receipt: the surviving commit itself still has no successful CI run. Re-run CI for that exact SHA and get it green; once that receipt exists, the code concern from my prior review is closed and the surviving chain satisfies the landing gate.

**Disposition:** functional review blocker resolved at `ef4687ee7a`; no new code finding. Landing sign-off remains withheld solely on the missing successful CI receipt for surviving commit `b9736f12f8fe`.

teknium1:
All five of your commits landed on `main` via #106563 (6530eb41adbb05cb6d690915a3d9a767150c7b23) with authorship preserved: dffc0fa2d54, 3db45bab6e7, 0efb5254206, 71818aaf522, 705bcdcf374. Closing as merged-via. Thanks @liuhao1024.

PR #98632, opened 2026-08-30
fix(tools): default search_files output_mode to files_only

### LINKED ISSUE #98626: search_files: output_mode defaults to content but should default to files_only — measured 3.4x token cost on unspecified calls
## Problem

`search_files`'s schema (`tools/file_tools.py:2804-2821`) defines `output_mode: content|files_only|count` with `default: "content"`. Both `SEARCH_FILES_SCHEMA` and `_handle_search_files` (`tools/file_tools.py:2867-2875`) fall back to this default identically.

`content` returns every matching line plus surrounding text; `files_only` returns just the matched file paths. The schema's own description already frames the two as "expensive, full detail" vs "cheap, just paths" — but the *default* points at the expensive one, and a model that omits the parameter (which is most of the time — see below) silently pays for detail it usually doesn't use, because the common next step is `read_file` on the file that matched anyway.

## Evidence (production telemetry, one instance, `~/.hermes/state.db`, 25 days / 74,843 tool calls)

- `target="content"` `search_files` calls that omitted `output_mode` (i.e. hit the default): **2,728 / 5,085 = 54%**
- Average payload size: **default (content) ≈ 2,957 B/call** vs **files_only ≈ 871 B/call** → **~3.4× larger**
- Reproduced in isolation, identical query, before/after the fix below (in-repo, not against production data): **446 B → 87 B (−80.5%)** on a 2-match search

## Proposed fix (patch attached, tested locally)

Change `SEARCH_FILES_SCHEMA`'s `output_mode` default from `"content"` to `"files_only"`, and the matching fallback in `_handle_search_files`. `target="files"` and explicit `output_mode="content"`/`"count"` requests are unaffected. Low-level `search_tool()` / `file_ops.search()` function defaults are left untouched — existing unit tests pin those independently at the lower layer, and changing them would be a wider blast radius than this fix needs.

Verified:
- 3 new regression tests at the schema/handler seam (schema default value; handler-omitted → files_only; handler-explicit-content → still content) — all pass
- Full related test suite (`test_file_tools.py` + 6 adjacent search/read-loop files, 152 tests) green — 2 pre-existing failures unrelated (macOS `/tmp` vs `/private/tmp` symlink resolution, present on `main` before this change too)

```diff
commit 109436e261c383bd3bc0d51ca314394850c79e36
Author: pitimon <9137713+pitimon@users.noreply.github.com>
Date:   Sun Aug 30 21:11:39 2026 +0700

    fix(search_files): default output_mode to files_only
    
    search_files defaulted output_mode to 'content', so any call that
    omitted it returned full matching lines instead of just paths.
    state.db telemetry across 25 days of real usage: 2,728 of 5,085
    content-target search_files calls (54%) never specified output_mode
    and fell into this default, averaging ~3.0KB/call vs files_only's
    ~0.9KB/call — the schema's own description already frames files_only
    as the cheap lookup and content as the expensive one, but the default
    pointed at the expensive path.
    
    A tool description is read once, a default is obeyed on every call —
    so the cheap mode has to be the default, not merely offered.
    
    Schema default + _handle_search_files fallback both changed to
    files_only; explicit output_mode requests (content/count) are
    unaffected. Low-level search_tool()/file_ops.search() defaults
    untouched since existing tests pin those independently.

diff --git a/tests/tools/test_file_tools.py b/tests/tools/test_file_tools.py
index 58fb91ac2..cd5219b57 100644
--- a/tests/tools/test_file_tools.py
+++ b/tests/tools/test_file_tools.py
@@ -305,6 +305,51 @@ class TestSearchHandler:
         assert "error" in result
 
 
+class TestSearchFilesDefaultOutputMode:
+    """A schema default is obeyed on every call while the description is
+    read once — the cheap mode has to be the default, not merely offered.
+    files_only costs a fraction of the tokens of full content (state.db:
+    content-mode search_files averaged ~3.0-3.6KB/call vs files_only's
+    ~0.9KB)."""
+
+    def test_schema_default_is_files_only(self):
+        from tools.file_tools import SEARCH_FILES_SCHEMA
+        assert (
+            SEARCH_FILES_SCHEMA["parameters"]["properties"]["output_mode"]["default"]
+            == "files_only"
+        )
+
+    @patch("tools.file_tools._get_file_ops")
+    def test_handler_omits_output_mode_defaults_to_files_only(self, mock_get):
+        mock_ops = MagicMock()
+        result_obj = MagicMock()
+        result_obj.to_dict.return_value = {"matches": ["file1.py"]}
+        mock_ops.search.return_value = result_obj
+        mock_get.return_value = mock_ops
+
+        from tools.file_tools import _handle_search_files
+        _handle_search_files({"pattern": "TODO"}, task_id="t-default-mode")
+
+        _, kwargs = mock_ops.search.call_args
+        assert kwargs.get("output_mode") == "files_only"
+
+    @patch("tools.file_tools._get_file_ops")
+    def test_handler_respects_explicit_content_request(self, mock_get):
+        mock_ops = MagicMock()
+        result_obj = MagicMock()
+        result_obj.to_dict.return_value = {"matches": ["file1.py:3:match"]}
+        mock_ops.search.return_value = result_obj
+        mock_get.return_value = mock_ops
+
+        from tools.file_tools import _handle_search_files
+        _handle_search_files(
+            {"pattern": "TODO", "output_mode": "content"}, task_id="t-explicit-mode"
+        )
+
+        _, kwargs = mock_ops.search.call_args
+        assert kwargs.get("output_mode") == "content"
+
+
 # ---------------------------------------------------------------------------
 # Windows MSYS path resolution (salvage of #50488 / #46995)
 # ---------------------------------------------------------------------------
diff --git a/tools/file_tools.py b/tools/file_tools.py
index ddcb06030..91e5cc83d 100644
--- a/tools/file_tools.py
+++ b/tools/file_tools.py
@@ -2803,7 +2803,7 @@ def _is_openai_family_main() -> bool:
 
 SEARCH_FILES_SCHEMA = {
     "name": "search_files",
-    "description": "Search file contents or find files by name. Use this instead of grep/rg/find/ls in terminal. Ripgrep-backed, faster than shell equivalents. On macOS, broad searches above the user home automatically skip TCC-protected folders (Desktop, Documents, Downloads, Library, Movies, Music, Pictures); target one directly when access is intentional.\n\nContent search (target='content'): Regex search inside files. Output modes: full matches with line numbers, file paths only, or match counts.\n\nFile search (target='files'): Find files by glob pattern (e.g., '*.py', '*config*'). Also use this instead of ls — results sorted by modification time.",
+    "description": "Search file contents or find files by name. Use this instead of grep/rg/find/ls in terminal. Ripgrep-backed, faster than shell equivalents. On macOS, broad searches above the user home automatically skip TCC-protected folders (Desktop, Documents, Downloads, Library, Movies, Music, Pictures); target one directly when access is intentional.\n\nContent search (target='content'): Regex search inside files. Output modes: full matches with line numbers, file paths only, or match counts. Defaults to file paths only, which costs a fraction of the tokens of full content — ask for output_mode='content' only when you need the matching lines themselves.\n\nFile search (target='files'): Find files by glob pattern (e.g., '*.py', '*config*'). Also use this instead of ls — results sorted by modification time.",
     "parameters": {
         "type": "object",
         "properties": {
@@ -2813,7 +2813,7 @@ SEARCH_FILES_SCHEMA = {
             "file_glob": {"type": "string", "description": "Filter files by pattern in grep mode (e.g., '*.py' to only search Python files)"},
             "limit": {"type": "integer", "description": "Maximum number of results to return (default: 50)", "default": 50},
             "offset": {"type": "integer", "description": "Skip first N results for pagination (default: 0)", "default": 0},
-            "output_mode": {"type": "string", "enum": ["content", "files_only", "count"], "description": "Output format for grep mode: 'content' shows matching lines with line numbers, 'files_only' lists file paths, 'count' shows match counts per file", "default": "content"},
+            "output_mode": {"type": "string", "enum": ["content", "files_only", "count"], "description": "Output format for grep mode: 'files_only' lists file paths (cheapest, default), 'content' shows matching lines with line numbers, 'count' shows match counts per file", "default": "files_only"},
             "context": {"type": "integer", "description": "Number of context lines before and after each match (grep mode only)", "default": 0}
         },
         "required": ["pattern"]
@@ -2872,7 +2872,7 @@ def _handle_search_files(args, **kw):
     return search_tool(
         pattern=args.get("pattern", ""), target=target, path=args.get("path", "."),
         file_glob=args.get("file_glob"), limit=args.get("limit", 50), offset=args.get("offset", 0),
-        output_mode=args.get("output_mode", "content"), context=args.get("context", 0), task_id=tid)
+        output_mode=args.get("output_mode") or "files_only", context=args.get("context", 0), task_id=tid)
 
 
 def _read_file_schema_overrides():
```

kokhlo:
Fixes #98626.

## What

`search_files` defaulted `output_mode` to `"content"`, so every grep-mode call that omitted the parameter returned full matching lines instead of just paths. The parameter's own description already frames `files_only` as the cheap lookup and `content` as the expensive one — but the default pointed at the expensive path, and the usual next step after a match is `read_file` on the matched file anyway, so the extra lines were paid for and then discarded.

A tool description is read once; a default is obeyed on every call. The cheap mode has to be the default, not merely offered.

## Change

- `SEARCH_FILES_SCHEMA` `output_mode` default: `"content"` → `"files_only"`, with the description reordered to lead with the cheapest mode.
- `_handle_search_files` fallback: `args.get("output_mode", "content")` → `args.get("output_mode") or "files_only"`.
- Explicit `output_mode` requests (`content`/`count`) and `target="files"` calls are unaffected.
- Low-level `search_tool()` / `file_ops.search()` function defaults are left untouched — existing tests pin those independently at that layer, and changing them would widen the blast radius beyond what this fix needs.

## Tests

Three regression tests at the schema/handler seam (`TestSearchFilesDefaultOutputMode`):
1. schema default value is `files_only`;
2. handler with omitted `output_mode` passes `files_only` downstream;
3. handler with explicit `content` still passes `content` downstream.

```
tests/tools/test_file_tools.py: 46 passed, 2 skipped
```
Two pre-existing failures on clean `origin/main` (mock assertion mismatches in `TestWriteFileHandler` / `TestPatchHandler`, unrelated to search) verified identical before/after via stash — not touched here.

### DIFF
--- a/tests/tools/test_file_tools.py
+++ b/tests/tools/test_file_tools.py
@@ -305,6 +305,49 @@ class TestSearchHandler:
         assert "error" in result
 
 
+class TestSearchFilesDefaultOutputMode:
+    """The schema default is obeyed on every call while the description is
+    read once — the cheap mode has to be the default, not merely offered.
+    Full content costs multiples of files_only per call."""
+
+    def test_schema_default_is_files_only(self):
+        from tools.file_tools import SEARCH_FILES_SCHEMA
+        assert (
+            SEARCH_FILES_SCHEMA["parameters"]["properties"]["output_mode"]["default"]
+            == "files_only"
+        )
+
+    @patch("tools.file_tools._get_file_ops")
+    def test_handler_omits_output_mode_defaults_to_files_only(self, mock_get):
+        mock_ops = MagicMock()
+        result_obj = MagicMock()
+        result_obj.to_dict.return_value = {"matches": ["file1.py"]}
+        mock_ops.search.return_value = result_obj
+        mock_get.return_value = mock_ops
+
+        from tools.file_tools import _handle_search_files
+        _handle_search_files({"pattern": "TODO"}, task_id="t-default-mode")
+
+        _, kwargs = mock_ops.search.call_args
+        assert kwargs.get("output_mode") == "files_only"
+
+    @patch("tools.file_tools._get_file_ops")
+    def test_handler_respects_explicit_content_request(self, mock_get):
+        mock_ops = MagicMock()
+        result_obj = MagicMock()
+        result_obj.to_dict.return_value = {"matches": ["file1.py:3:match"]}
+        mock_ops.search.return_value = result_obj
+        mock_get.return_value = mock_ops
+
+        from tools.file_tools import _handle_search_files
+        _handle_search_files(
+            {"pattern": "TODO", "output_mode": "content"}, task_id="t-explicit-mode"
+        )
+
+        _, kwargs = mock_ops.search.call_args
+        assert kwargs.get("output_mode") == "content"
+
+
 # ---------------------------------------------------------------------------
 # Windows MSYS path resolution (salvage of #50488 / #46995)
 # ---------------------------------------------------------------------------
--- a/tools/file_tools.py
+++ b/tools/file_tools.py
@@ -2744,7 +2744,7 @@ SEARCH_FILES_SCHEMA = {
             "file_glob": {"type": "string", "description": "Filter files by pattern in grep mode (e.g., '*.py' to only search Python files)"},
             "limit": {"type": "integer", "description": "Maximum number of results to return (default: 50)", "default": 50},
             "offset": {"type": "integer", "description": "Skip first N results for pagination (default: 0)", "default": 0},
-            "output_mode": {"type": "string", "enum": ["content", "files_only", "count"], "description": "Output format for grep mode: 'content' shows matching lines with line numbers, 'files_only' lists file paths, 'count' shows match counts per file", "default": "content"},
+            "output_mode": {"type": "string", "enum": ["content", "files_only", "count"], "description": "Output format for grep mode: 'files_only' lists file paths (cheapest, default), 'content' shows matching lines with line numbers, 'count' shows match counts per file", "default": "files_only"},
             "context": {"type": "integer", "description": "Number of context lines before and after each match (grep mode only)", "default": 0}
         },
         "required": ["pattern"]
@@ -2803,7 +2803,7 @@ def _handle_search_files(args, **kw):
     return search_tool(
         pattern=args.get("pattern", ""), target=target, path=args.get("path", "."),
         file_glob=args.get("file_glob"), limit=args.get("limit", 50), offset=args.get("offset", 0),
-        output_mode=args.get("output_mode", "content"), context=args.get("context", 0), task_id=tid)
+        output_mode=args.get("output_mode") or "files_only", context=args.get("context", 0), task_id=tid)
 
 
 registry.register(name="read_file", toolset="file", schema=READ_FILE_SCHEMA, handler=_handle_read_file, check_fn=_check_file_reqs, emoji="📖", max_result_size_chars=100_000)


webtecnica (commented):
Verified this fix against current `main` — it's correct and ready. Details from a full RED-GREEN run on a worktree of upstream/main:

**What was verified**
- `SEARCH_FILES_SCHEMA` `output_mode` default flipped `content` → `files_only` (line 2816 on current main).
- `_handle_search_files` fallback flipped to `args.get("output_mode") or "files_only"` (line 2875) — so an explicit `output_mode="content"`/`"count"` still wins, only the omitted case changes.
- 3 regression tests (`TestSearchFilesDefaultOutputMode`) run clean: schema default assertion, handler-omitted → `files_only`, handler-explicit-`content` → still `content`.
- RED confirmed: stashing the production change while keeping the new tests yields 2 failures (schema default + handler default); explicit-content test passes as expected. GREEN with the fix: `tests/tools/test_file_tools.py` + the 6 adjacent search/read-loop test files = **96 passed, 2 skipped** (the 2 skips are the pre-existing macOS `/tmp` vs `/private/tmp` symlink cases).
- No dependents of the old default found: low-level `search_tool()`/`file_ops.search()` defaults are deliberately untouched (existing unit tests pin them), and `tests/tools/test_search_error_guard.py` + `tools/code_execution_tool.py` don't rely on the handler/schema default.

**One thing to consider before merge**
The PR branch is based on a `main` from ~2,091 commits ago (merge-base `4a5b6dd451`). GitHub reports it MERGEABLE (3-way), but a rebase onto current `main` would be cleaner and de-risk any CI conflicts — the fix itself ports with only a ~5-line offset.

**Optional (matches the issue's own proposed patch)**
The issue body's patch also rewrote the *top-level* `SEARCH_FILES_SCHEMA` description with a token-cost note ("Defaults to file paths only, which costs a fraction of the tokens of full content…"). This PR only updates the `output_mode` property description. The property-level change is the one the model actually reads per-parameter, so the top-level rewrite is a nice-to-have, not a blocker.

Thanks for picking this up — the 3.4× payload asymmetry on omitted-parameter calls (54% of `content`-target searches in the issue's telemetry) makes this a genuinely good perf win.

kshitijk4poor:
**Triage findings (perf/P2 sweep)** — comment only; this is a product-behaviour call for a maintainer, not something the sweep decides.

What the change does: flips the `search_files` default `output_mode` from `content` to `files_only` in both the schema default and the handler fallback (`args.get("output_mode") or "files_only"`, which also maps an explicit empty string to `files_only`).

What breaks / changes for existing callers relying on the default:
- Every skill and prompt that calls `search_files(pattern=…)` expecting matching lines now gets file paths only — a grep of `skills/` finds 40 such `search_files(` references with no `output_mode`, plus whatever habits deployed models have formed.
- The `context=` parameter becomes a silent no-op under the new default (it only applies in content mode).
- No perf benchmark applies: the saving is output tokens per call, which depends entirely on the query.

Tests: 47 pass; the 2 failures in `tests/tools/test_file_tools.py` (`/tmp` vs `/private/tmp`) are **pre-existing on main** on macOS, not caused by this PR. Applies with a one-hunk context conflict against 1cb3ab6173.

kshitijk4poor:
Follow-up measurement, to ground the maintainer decision (leaving this open — it is a product-behaviour call, not something the perf sweep decides).

I profiled where the bytes in a default (`content`) `search_files` response actually go, across 10 realistic queries against this repo on current `main` (39 KB total; main already has the path-grouped densification ≥5 matches and #95940's giant-line bound):

| component | share |
|---|---|
| matched-line text (after the `<line>: ` prefix, excluding indentation) | ~55% |
| file paths | 20% |
| leading indentation of matched lines | 7% |
| line tails beyond 120 chars | 7% |
| JSON / `matches_format` overhead | 7% |
| line numbers | 5% |

Same queries, `content` vs `files_only`: 1.1×–7× (median ~2.3×), consistent with the 3.4× in #98626's telemetry.

What that means for the options:
- **Make `content` mode cheaper while keeping lines** — the only fat is indentation + long tails + format boilerplate, ~15–20% total, and stripping indentation loses information in Python. This cannot recover the 3.4×; the bytes *are* the matching lines. I had planned an alternative PR along these lines and am not opening it, because the measurement refutes its premise.
- **Flip the default (this PR)** — recovers the full saving, but changes the tool's contract: `search_files("error string")` stops answering *where* in the file, `context=` becomes a silent no-op under the default, the handler's `or "files_only"` also remaps an explicit `""`, and 38 skill call sites + the system prompt's own guidance ("file contents, line counts → search_files") assume lines. Whether the per-call saving is worth that is the call.
- **Middle ground worth considering** — keep `content` as the default but change what the model is *told*: reorder the schema description to lead with `files_only` as the cheap discovery step and make `content` the explicit "I need the lines" mode. That changes model behaviour on the 54% default-omitting calls without breaking any caller that relies on the default. If a maintainer prefers this, it's a description-only change I'm happy to send.

Method: `search_tool()` called directly with real imports, absolute repo paths, `HERMES_HOME` isolated; byte attribution by parsing `matches_text`.

PR #98633, opened 2026-08-30
fix(context): compact under message-count pressure

anhtahaylove:
## Summary
- treat hard gateway message-row limits independently from token pressure
- force a bounded recent tail once a transcript reaches 600 rows
- keep the manual `/compress` preflight consistent with the actual compression boundary
- add a regression test covering an 803-row transcript and assert the committed result falls below 600 rows

## Why
Some OpenAI-compatible gateways reject chat histories at 800 message rows even when the request still fits the model token window. Token-based tail protection could therefore produce an in-place `800 -> 800` compaction and subsequent `413 Cannot compress further` failures.

## Verification
- `uv run pytest tests/agent/test_message_count_pressure.py tests/agent/test_protected_tail_pressure_61932.py tests/agent/test_compression_progress.py tests/agent/test_compression_attempt_telemetry.py -q` (11 passed)
- `uv run ruff check agent/context_compressor.py tests/agent/test_message_count_pressure.py`
- `git diff --check`

### DIFF
--- a/agent/context_compressor.py
+++ b/agent/context_compressor.py
@@ -972,6 +972,13 @@ _MAX_TAIL_MESSAGE_FLOOR = 8
 # Skip the LLM call when the compressible middle is below this fraction of the
 # threshold (and a prior ineffectiveness strike exists); dropping alone suffices.
 # See #60451.
+# A long tool/background-heavy transcript can fit the tail token budget while still
+# exceeding a gateway's hard row limit, leaving no middle window and making every
+# 413 retry a no-op. Past this many rows, force a boundary that keeps the newest
+# _MESSAGE_COUNT_PRESSURE_TAIL rows and summarizes everything older.
+_MESSAGE_COUNT_PRESSURE_THRESHOLD = 600
+_MESSAGE_COUNT_PRESSURE_TAIL = 200
+
 _FEASIBILITY_SKIP_MIDDLE_FRACTION = 0.10
 # Under pressure, demote large tool outputs even inside the protected region but
 # keep this many trailing messages verbatim.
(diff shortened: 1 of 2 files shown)


kokhlo:
Tracing the pressure path end to end — the forced boundary at 600 messages is the right lever, but there is a case where it still no-ops:

the feasibility-skip pre-LLM gate (`_FEASIBILITY_SKIP_MIDDLE_FRACTION = 0.10`) can defeat the forced boundary when the middle window is token-light. Consider a 600+ row transcript of lightweight background events (~5 tokens each): the forced boundary protects the tail 200 rows, but the ~200-row middle might weigh ~1000 tokens. Against a 272K `threshold_tokens` that is 0.37% — far below the 10% feasibility floor — so feasibility-skip discards those rows with no summary despite message-count pressure having fired.

If that is an accepted trade-off (tiny middle windows are rarely worth an LLM call), a comment stating it explicitly would preempt the question; if not, treating `message_count_pressure` as a force-bypass of the feasibility check (or asserting `feasibility_skip=False` when `telemetry["message_count_pressure"]` is set) closes it. Not blocking — the main pressure logic reads correct.

anhtahaylove:
Good catch — I agreed with the semantic invariant and fixed it in `b27fd27303`.

Message-count pressure now marks `telemetry["message_count_pressure"]` as soon as the 600-row threshold is reached, even when boundary alignment does not have to extend `compress_end`. The feasibility gate explicitly bypasses its 10% floor whenever that flag is set, so a token-light middle window can no longer turn row pressure into a silent no-op.

I also added a regression with 600 lightweight messages, `_ineffective_compression_count = 1`, and a middle window below the normal feasibility floor. It asserts `message_count_pressure=True`, `feasibility_skip=False`, and that the summary client is actually invoked.

Verified against current `main`: `uv run pytest -q tests/agent/test_message_count_pressure.py` — 2 passed.

anhtahaylove:
Gentle nudge: this one is **already through CI** and needs no further compute to evaluate.

At head `39d7533589`: `check-runs total_count = 37`, all workflow conclusions `success`, and the API reports `mergeable = true` / `mergeable_state = clean` — no conflicts with `main`.

It's a small diff (2 files, +104/-0) and has been open since Aug 30, so it's purely waiting on a reviewer rather than on the fork-PR workflow gate that's holding most of my other PRs.

I won't push anything further here — a new push would discard the green runs and send it back to the approval queue. Happy to rebase or close it if it's no longer wanted.

PR #98638, opened 2026-08-30
fix(gateway): keep wrapped containers online after restart collisions

### LINKED ISSUE #98625: [Bug]: gateway auto-restart collides with stale lock, gives up permanently after 5 attempts
### Bug Description

When the gateway process receives SIGTERM (in our case sent by the container's own supervisor, parent_cmdline=/app/u4s-hermes-agent, not by a user action), the built-in auto-restart loop tries to relaunch the gateway. Every retry immediately fails with "Another gateway instance is already running (PID N)" because the previous process/lock hasn't been released yet. After 5 failed attempts (backoff 0s/5s/10s/15s/20s, about 50s total) it logs "agent exceeded 5 restart attempts, giving up" and never tries again. The gateway then stays down indefinitely (about 38 hours in our case) until a human notices and manually restarts the whole container/app, which clears the stale lock.

This defeats the purpose of the auto-restart mechanism: a transient SIGTERM turns into a permanent, silent outage.

### Steps to Reproduce

Hard to trigger deliberately, since the initial SIGTERM in our case came from the container's own supervisor, not from a user action. The failure pattern reproduces reliably whenever a SIGTERM/crash happens while the previous gateway process/lock is still being released at the moment the supervisor's restart loop fires:

1. Gateway process running normally (hermes gateway run, launched by the container's own supervisor).
2. 2. Something sends SIGTERM to the gateway (source unclear in our case; not OOM, per suspected_oom=False in the lifecycle ledger log).
3. 3. Supervisor's built-in restart loop immediately tries to relaunch the gateway.
4. 4. Relaunch fails: "Another gateway instance is already running (PID N). Use 'hermes gateway restart' to replace it, or 'hermes gateway stop' first. Or use 'hermes gateway run --replace' to auto-replace." The previous process/lock has not released yet.
5. 5. Retries at 5s/10s/15s/20s backoff, all fail the same way.
6. 6. After 5 attempts: "agent exceeded 5 restart attempts, giving up". Gateway stays down with no further recovery attempt.

### Expected Behavior

The auto-restart loop should recover from this stale-lock condition instead of giving up permanently. For example, retry with `hermes gateway run --replace` after the first (or after N) plain-retry failures, or wait/verify the old PID has actually exited before retrying. A transient SIGTERM should not be able to cause a permanent, unattended outage.

### Actual Behavior

Exact log excerpt from the Hostinger "Hermes Agent" managed-app panel (Panel > Historico), timestamps in the app's local log format:

time=23:34:55 level=INFO msg="[agent] 2026-08-28 23:34:55,034 WARNING gateway.run: Shutdown context: signal=SIGTERM under_systemd=no parent_pid=7 parent_name=u4s-hermes-agen loadavg_1m=15.31 parent_cmdline=/app/u4s-hermes-agent"
time=23:34:56 level=WARN msg="agent exited with code 0"
time=23:34:56 level=INFO msg="Restarting agent in 0s (attempt 0/5)"
time=23:34:56 level=WARN msg="agent exited with error: signal: terminated"
time=23:34:56 level=INFO msg="Restarting agent in 5s (attempt 1/5)"
time=23:35:02 level=INFO msg="[agent] Another gateway instance is already running (PID 300)."
time=23:35:02 level=INFO msg="[agent] Use 'hermes gateway restart' to replace it,"
time=23:35:02 level=INFO msg="[agent] or 'hermes gateway stop' first."
time=23:35:02 level=INFO msg="[agent] Or use 'hermes gateway run --replace' to auto-replace."
time=23:35:02 level=WARN msg="agent exited with error: exit status 1"
time=23:35:02 level=INFO msg="Restarting agent in 10s (attempt 2/5)"
... (attempts 2, 3, 4 repeat the same "Another gateway instance is already running (PID 300)" failure, backoff 10s/15s/20s) ...
time=23:35:49 level=WARN msg="agent exited with error: exit status 1"
time=23:35:49 level=ERROR msg="agent exceeded 5 restart attempts, giving up"

No further [agent] log lines appear until a human manually restarts the app roughly 38 hours later. Only [webui] health-check lines continue during that gap.

### Affected Component

Gateway (Telegram/Discord/Slack/WhatsApp)

Telegram

### Debug Report

```shell
I don't have a `hermes debug share` output for this: the gateway had already been down for ~38h when I got involved, and by the time it was accessible again the incident window had passed and the CLI session had reset. Instead of a debug-share paste, I'm attaching the exact raw agent/gateway log excerpt from the Hostinger managed-app panel's own log viewer (Panel > Historico), pasted verbatim in the Actual Behavior field above, since it captures this specific incident (including the SIGTERM, the 5 retry attempts, and the PID collision) better than a debug-share snapshot taken after the fact would.
```

### Operating System

Linux (Docker container, Hostinger "Hermes Agent" managed app; under_systemd=no per gateway.run shutdown log)

### Python Version

_No response_

### Hermes Version

_No response_

```shell

```

I haven't inspected the source, so no file/line pointer, just a behavioral hypothesis from the logs: the supervisor's restart loop appears to always invoke plain `hermes gateway run` on every attempt, with a fixed backoff (0s/5s/10s/15s/20s) and a hard cap of 5 tries. If the previous process/lock (PID file or similar) is not released within that ~50s window, every attempt hits the same "Another gateway instance is already running" guard, so the loop is guaranteed to fail identically 5 times and then stop trying forever, rather than backing off further or switching strategy.

### Proposed Fix (optional)

A few options, roughly in order of how surgical they'd be:
1. Have the restart loop fall back to `hermes gateway run --replace` starting from attempt 2 (or after the first plain-run failure), since that flag already exists to handle exactly this collision.
2. 2. Before retrying, actively check whether the PID reported in "Another gateway instance is already running (PID N)" is still alive; if not, treat the lock as stale and clear it instead of waiting/retrying blindly.
3. 3. Don't give up permanently after a fixed attempt count. At minimum, keep retrying at a longer interval (e.g. every few minutes) indefinitely, and/or surface a clear persistent alert/log line so a human notices sooner than "silently down until someone happens to check".

fangliquanflq:
## What does this PR do?

Wrapped container runtimes now recover when a gateway restart overlaps the previous process releasing its runtime lock. Without this change, every bounded supervisor retry can hit the same "Another gateway instance is already running" refusal and leave messaging offline until an operator restarts the app.

### Symptom

A platform-managed container sends SIGTERM to the gateway and immediately relaunches `gateway run`. If the previous gateway still owns the runtime lock, each relaunch exits with the duplicate-instance error until the external supervisor exhausts its retry budget.

### Impact

Affected wrapped container deployments can remain offline indefinitely after a transient restart collision. The reported incident required a manual restart after approximately 38 hours without gateway service.

### Bug Cause

**Trigger:** `docker/entrypoint-dispatch.sh:21-25` / the non-PID-1 direct-bootstrap fallback.

**Causal chain:**
1. A runtime that wraps the image entrypoint owns PID 1, so Hermes skips s6 and directly launches the requested `gateway run` command.
2. The external supervisor restarts that command before the previous gateway has released its live PID/runtime lock.
3. The new process follows the ordinary foreground duplicate guard, exits non-zero, and every bounded retry repeats the same refusal.

**Why it is wrong:** The wrapped runtime is the authoritative owner of this process slot, but the fallback did not preserve that ownership signal or enable the existing identity-checked takeover path.

**Working sibling / contrast:** Standard systemd, launchd, and s6 gateway services already launch with `--replace`, so their authoritative child can safely reap a stale holder and bind.

**Ruled out:** Globally weakening the duplicate-instance guard is unsafe because an interactive `hermes gateway run` must not kill a legitimate running gateway. The change is gated on explicit external-supervisor ownership.

### Fix

The non-PID-1 Docker fallback now exports the existing external-supervisor marker before bootstrap. Gateway command dispatch treats that explicit marker as authoritative restart ownership and enables the existing `--replace` behavior, which validates process identity, performs SIGTERM/SIGKILL takeover with confirmation, and cleans scoped locks. Ordinary foreground runs retain the duplicate-instance refusal.

## Related Issue

Closes #98625

## Type of Change

## Changes Made

- `docker/entrypoint-dispatch.sh` - mark the direct-bootstrap wrapped-runtime path as externally supervised.
- `hermes_cli/gateway.py` - enable the hardened takeover path for explicitly marked external supervisors.
- `tests/hermes_cli/test_entrypoint_dispatch_fallback.py` - execute the real dispatcher fallback and verify marker propagation.
- `tests/hermes_cli/test_gateway_external_supervisor.py` - verify supervised takeover and preserve ordinary foreground refusal.

## How to Test

1. Execute the non-PID-1 dispatcher with isolated stage2/wrapper probes and verify the wrapper receives `HERMES_GATEWAY_EXTERNAL_SUPERVISOR=1`.
2. Dispatch `gateway run` with that marker and verify `run_gateway(..., replace=True)`.
3. Dispatch an ordinary foreground run without the marker and verify `replace=False`.
4. Run the relevant automated tests:

```bash
scripts/run_tests.sh tests/hermes_cli/test_gateway.py -q
scripts/run_tests.sh tests/hermes_cli/test_gateway_external_supervisor.py -q
scripts/run_tests.sh tests/hermes_cli/test_entrypoint_dispatch_fallback.py -q
scripts/run_tests.sh tests/gateway/test_restart_service_detection.py -q
```

All 46 relevant tests passed locally.

### Code

- [x] I've run the repo's test entry on the relevant tests and all tests pass
- [x] I've tested on Windows 11 with the portable shell dispatcher exercised through Git Bash

- [x] Documentation update is N/A; behavior is internal recovery hardening
- [x] Cross-platform impact considered; takeover remains gated on the existing supervisor marker
- [x] Tool descriptions/schemas update is N/A; no model tool behavior changed

## Screenshots / Logs

N/A; automated behavior tests cover the restart ownership contract.

### DIFF
--- a/docker/entrypoint-dispatch.sh
+++ b/docker/entrypoint-dispatch.sh
@@ -24,11 +24,15 @@ WRAPPER="${HERMES_ENTRYPOINT_WRAPPER:-/opt/hermes/docker/main-wrapper.sh}"
 echo "[hermes] WARNING: container entrypoint is not PID 1; skipping s6-overlay /init and falling back to direct bootstrap. Supervised services are unavailable in this runtime, but the requested command will still run." >&2
 # /init normally seeds PATH with s6's helpers; the non-PID-1 fallback skips it.
 export PATH="/command:/package/admin/s6/command:${PATH}"
-# The platform wrapper owns this process slot just as systemd/launchd/s6 own
-# theirs. Mark that ownership so a restarted ``gateway run`` may use the
-# hardened takeover path when the previous process has not released its
-# runtime lock yet, rather than repeating the same duplicate-instance failure
-# until the platform exhausts its retry budget (#98625).
-export HERMES_GATEWAY_EXTERNAL_SUPERVISOR=1
 "$STAGE2"
+
+# The platform wrapper owns this process slot just as systemd/launchd/s6 own
+# theirs. Mark only the gateway process that the wrapper launches directly;
+# descendants inherit the environment marker, so takeover authorization must
+# remain on the explicit CLI flag rather than the ambient environment.
+if [ "${1:-}" = "gateway" ] && [ "${2:-}" = "run" ]; then
+    export HERMES_GATEWAY_EXTERNAL_SUPERVISOR=1
+    exec "$WRAPPER" "$@" --external-supervisor
+fi
+
 exec "$WRAPPER" "$@"
--- a/docker/entrypoint-dispatch.sh
+++ b/docker/entrypoint-dispatch.sh
@@ -27,40 +27,8 @@ export PATH="/command:/package/admin/s6/command:${PATH}"
 "$STAGE2"
 
 # The platform wrapper owns this process slot just as systemd/launchd/s6 own
-# theirs. Mark only a gateway process that the wrapper launches directly;
-# descendants inherit environment markers, so takeover authorization remains
-# on the explicit CLI flag rather than ambient state.
-if [ "${1:-}" = "gateway" ]; then
-    if [ "$#" -eq 1 ]; then
-        exec "$WRAPPER" gateway run --external-supervisor
-    elif [ "${2:-}" = "run" ]; then
-        exec "$WRAPPER" "$@" --external-supervisor
-    fi
-elif [ "${1:-}" = "hermes" ] && [ "${2:-}" = "gateway" ]; then
-    if [ "$#" -eq 2 ]; then
-        exec "$WRAPPER" hermes gateway run --external-supervisor
-    elif [ "${3:-}" = "run" ]; then
-        exec "$WRAPPER" "$@" --external-supervisor
-    fi
-elif { [ "${1:-}" = "--profile" ] || [ "${1:-}" = "-p" ]; } && \
-     [ "${3:-}" = "gateway" ]; then
-    if [ "$#" -eq 3 ]; then
-        exec "$WRAPPER" "$@" run --external-supervisor
-    elif [ "${4:-}" = "run" ]; then
-        exec "$WRAPPER" "$@" --external-supervisor
-    fi
-else
-    case "${1:-}" in
-        --profile=*)
-            if [ "${2:-}" = "gateway" ]; then
-                if [ "$#" -eq 2 ]; then
-                    exec "$WRAPPER" "$@" run --external-supervisor
-                elif [ "${3:-}" = "run" ]; then
-                    exec "$WRAPPER" "$@" --external-supervisor
-                fi
-            fi
-            ;;
-    esac
-fi
-
+# theirs. exec preserves this PID through main-wrapper and privilege drop, so
+# the gateway can authorize this exact slot without trusting an ambient marker
+# inherited by later child processes.
+export HERMES_GATEWAY_EXTERNAL_SUPERVISOR_PID="$$"
 exec "$WRAPPER" "$@"
--- a/docker/entrypoint-dispatch.sh
+++ b/docker/entrypoint-dispatch.sh
@@ -27,12 +27,40 @@ export PATH="/command:/package/admin/s6/command:${PATH}"
 "$STAGE2"
 
 # The platform wrapper owns this process slot just as systemd/launchd/s6 own
-# theirs. Mark only the gateway process that the wrapper launches directly;
-# descendants inherit the environment marker, so takeover authorization must
-# remain on the explicit CLI flag rather than the ambient environment.
-if [ "${1:-}" = "gateway" ] && [ "${2:-}" = "run" ]; then
-    export HERMES_GATEWAY_EXTERNAL_SUPERVISOR=1
-    exec "$WRAPPER" "$@" --external-supervisor
+# theirs. Mark only a gateway process that the wrapper launches directly;
+# descendants inherit environment markers, so takeover authorization remains
+# on the explicit CLI flag rather than ambient state.
+if [ "${1:-}" = "gateway" ]; then
+    if [ "$#" -eq 1 ]; then
+        exec "$WRAPPER" gateway run --external-supervisor
+    elif [ "${2:-}" = "run" ]; then
+        exec "$WRAPPER" "$@" --external-supervisor
+    fi
+elif [ "${1:-}" = "hermes" ] && [ "${2:-}" = "gateway" ]; then
+    if [ "$#" -eq 2 ]; then
+        exec "$WRAPPER" hermes gateway run --external-supervisor
+    elif [ "${3:-}" = "run" ]; then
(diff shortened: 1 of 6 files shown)


Enough1122:

Fixes wrapped-container restart authority correctly by binding supervisor ownership to the exec PID slot rather than an inheritable env flag.

- **PID-bound marker (`docker/entrypoint-dispatch.sh:9+`, `gateway/restart.py:52+`):** `HERMES_GATEWAY_EXTERNAL_SUPERVISOR_PID` set to `$$` before `exec` preserves identity through `main-wrapper` and privilege drop. Prevents children inheriting destructive restart authority.
- **Detection split (`gateway/control_socket.py:33+`, `hermes_cli/gateway.py:73+`):** `_detect_supervisor` now checks env PID == `getpid()`, and `_gateway_process_has_bound_external_supervisor` verifies via `psutil.Process(pid).environ()` for update clients querying other PIDs. Handles PID 1 guard and psutil failures gracefully.
- **Update path (`hermes_cli/gateway.py:101+`):** `_prepare_profile_gateway_update_restart` treats both `--external-supervisor` arg and PID-bound case as `external-supervisor`, avoiding racing detached watcher (`launch_detached...` not called). Matches systemd/launchd/s6 existing `--replace` semantics by forcing `replace=True` for bound slot.
- **Tests:** `test_entrypoint_dispatch_fallback.py` verifies argv preservation and PID equality; `test_gateway_external_supervisor.py` covers flag-takeover, PID-takeover, inherited-marker negative, and plain run — solid matrix.

Non-blocking:
- `psutil` is optional import; exception swallows to `False`. If psutil absent, PID-bound detection degrades to argv-only — document or guard CI to ensure psutil available in production image.

fangliquanflq:
Thanks for the review. I checked the non-blocking `psutil` point and no code or documentation change is needed:

- `psutil==7.2.2` is a pinned core dependency in `pyproject.toml`, not an optional production dependency. The PR does not change either `pyproject.toml` or `uv.lock`, and current `main` has the identical dependency declaration.
- The local import and fail-closed exception handling in `_gateway_process_has_bound_external_supervisor()` cover process-exit/permission races as well as broken environments; they do not define a supported psutil-free install. If inspection fails, the code does not grant external-supervisor ownership.
- The PR changes only the dispatcher, supervisor detection/restart paths, and their tests. The three failing CI tests are in unchanged goal/compression files and all pass through the canonical isolated runner locally.

Validation:
- `scripts/run_tests.sh tests/hermes_cli/test_gateway.py tests/hermes_cli/test_gateway_external_supervisor.py tests/hermes_cli/test_entrypoint_dispatch_fallback.py tests/gateway/test_restart_service_detection.py -q` — 60 passed, 5 skipped.
- `scripts/run_tests.sh tests/gateway/test_goal_continuation_drain.py tests/agent/test_compression_worker_isolation_76354.py tests/gateway/test_goal_resume_restart.py -q` — 9 passed.

PR #36689, opened 2026-06-01
feat(gateway): previous-session bridge for auto-reset continuity

avi-shual:
## What does this PR do?

Adds a **Previous-Session Bridge** to the messaging gateway: when a 1-on-1 session auto-resets due to idle/daily timeout and the user sends a follow-up message, the new session's system prompt is seeded with a short tail of the prior conversation (last few user/assistant text exchanges, no tool calls). The agent has continuity with what the user last saw and doesn't lose the thread mid-thought.

**Problem this solves:** Today, idle/daily resets are a hard cut. A user who steps away for 8 hours and resumes with "yeah, do that" gets a fresh agent that can't recover the antecedent. Persistent memory holds the durable facts, but not the immediate conversational context. This bridge fills that gap with a small (~1000-token), bounded prompt addendum that's also safe to ship by default.

**Why this design:**
- *Auto-reset only* — explicit `/reset` / `/new` is always a clean slate. We respect the intent.
- *Last-N text exchanges* — bounded by both `max_exchanges` and `max_chars`. No tool calls, no reasoning traces — just what the user actually saw.
- *Append to system prompt*, not message history — preserves prompt-cache hit rate on the durable prefix.
- *Always-on privacy guards* (not configurable knobs):
  - Skipped for **shared multi-user sessions** (group chats / channels) — no cross-user leak.
  - Skipped when **`redact_pii` is enabled** for the platform — the redaction pipeline doesn't run over the historical tail, so we skip rather than leak.
  - Skipped on **`force_new=True`** (explicit `/reset`, `/new`).
- *Best-effort resolution* — if the prior session_id can't be loaded (deleted, DB error), the bridge silently skips and the new session starts fresh.

## Related Issue

No existing issue — this is a small UX improvement surfaced during dogfooding.

Fixes #N/A

## Type of Change

- [x] ✨ New feature (non-breaking change that adds functionality)

## Changes Made

- `gateway/session.py` — adds `SessionEntry.previous_session_id`, `should_bridge_previous_session()` predicate, `render_previous_session_tail()` renderer, and `build_session_context_prompt(previous_session_tail=...)` integration
- `gateway/config.py` — new `PreviousSessionBridge` dataclass (`enabled`, `max_exchanges`, `max_chars`); wired into `GatewayConfig` and `cli-config.yaml` loader so `previous_session_bridge:` top-level key is honored alongside `session_reset:`
[...]

PR #13536, opened 2026-04-21
fix(gateway): persist parent session ID on auto-reset

Tranquil-Flow:
## What does this PR do?

When the gateway auto-resets a session (e.g. after suspension), the new session had no link back to the previous one. This makes it impossible to trace session lineage for debugging or analytics.

This adds a `parent_session_id` field to `SessionEntry` that gets set to the old session's ID during auto-reset. The field is:
- Persisted through `to_dict()`/`from_dict()` serialization
- Forwarded to the SQLite `create_session()` call for the session graph
- Defaults to `None` for fresh sessions (no reset)
- Backward-compatible with existing session data (missing field defaults to `None`)

## Related Issue

Closes #12857

## Type of Change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [x] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made

- `SessionEntry` — new `parent_session_id` field, persisted through `to_dict()` / `from_dict()` serialization, defaults to `None` for fresh sessions
- Auto-reset path sets `parent_session_id` to the old session's ID before swapping in the new entry
- SQLite `create_session()` call now forwards `parent_session_id` for the session graph
- `tests/gateway/test_session_parent_id.py` — 9 tests covering field acceptance, default None, serialization roundtrip, missing/null handling in `from_dict`, auto-reset propagation, fresh session behavior, and SQLite forwarding

## How to Test

1. ```bash
   pytest tests/gateway/test_session_parent_id.py -v
   ```
2. 9 tests pass covering field acceptance, default None, serialization roundtrip, missing/null handling in `from_dict`, auto-reset propagation, fresh session behavior, and SQLite forwarding

## Checklist

### Code

- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
- [x] My PR contains **only** changes related to this fix/feature (no unrelated commits)
[...]

PR #98640, opened 2026-08-30
fix(gateway): retain original auto-reset lineage in continuity hints

smicbee:
## What does this PR do?

Extends the existing Slack/Discord auto-reset continuity hint so a fresh session receives both:

- the immediate predecessor session ID, and
- the original/root session ID of the contiguous auto-reset lineage.

The gateway already records `parent_session_id` for reset continuations and points the model at the immediate predecessor. After multiple daily/idle resets, however, the model can only walk one hop and may recover an intermediate session that no longer contains the original antecedent.

The new session now walks the existing durable parent chain once at reset time, stores the root ID in the routing entry's lightweight metadata, and includes it in the one-shot turn sidecar note. This preserves the byte-stable system prompt and keeps historical content untrusted: no prior transcript text is elevated into the system prompt.

Explicit `/new` and `/reset` boundaries are unchanged because the root is computed only for `was_auto_reset` entries.

## Problem reproduced

In a long-lived Discord thread with multiple automatic resets:

1. Session A contains the original topic.
2. Daily reset creates session B, whose immediate predecessor is A.
3. Another reset creates session C, whose immediate predecessor is B.
4. The first message in C references the original topic.

Before this change, the note only identifies B. If B is sparse or has drifted, `session_search` cannot reliably recover A without guessing. After this change, the note identifies both B and A and explicitly directs the model to inspect the original lineage session when the request points back to the original topic.

## Implementation

- Adds a bounded, cycle-safe parent walker using the existing `SessionDB.get_session()` records.
- Stops at session-key boundaries, malformed rows, missing parents, or 256 hops.
- Stores `continuity_root_session_id` in `SessionEntry.metadata`, which already persists through gateway restarts.
- Extends `build_channel_continuity_note()` without adding model tools, config keys, transcript injection, or extra lookups on normal turns.
- Adds Discord-specific and multi-reset regression coverage.

## Duplicate sweep

- #36689 is related but injects prior transcript text into the system prompt and is currently stale against main; this PR only carries deterministic session IDs through the existing safe sidecar mechanism.
- #13536 addressed parent-session persistence; current main already contains durable reset parent links, which this PR consumes.

## Verification

```text
18 passed in 6.30s
All checks passed!  # ruff
```

Commands:

```bash
python -m pytest \
  tests/gateway/test_channel_continuity_hint.py \
  tests/gateway/test_session_store_runtime_stale_guard.py \
  tests/gateway/test_10710_auto_reset_evicts_cached_agent.py \
  -q -o 'addopts='
python -m ruff check gateway/session.py tests/gateway/test_channel_continuity_hint.py
git diff --check origin/main...HEAD
```

The regression test was also sabotage-checked against the pre-fix `gateway/session.py`: 2 lineage assertions failed, then passed again after restoring the fix.

## Risk

Low and bounded:

- one parent-chain read only when creating an auto-reset replacement session;
- no behavior change for manual resets, non-Slack/Discord platforms, activity-free resets, or ordinary turns;
- no prior message content is copied into higher-authority context.

### DIFF
--- a/gateway/session.py
+++ b/gateway/session.py
@@ -1035,12 +1035,24 @@ def build_channel_continuity_note(
     if not prev:
         return None
 
+    root = (getattr(entry, "metadata", None) or {}).get(
+        "continuity_root_session_id"
+    )
+    root_hint = ""
+    if root and root != prev:
+        root_hint = (
+            f" The original session in this auto-reset lineage is "
+            f"session_id: {root}."
+        )
+
     where = "thread" if source.thread_id else "channel"
     return (
         f"[System note: This {where} had an earlier Hermes session "
-        f"(session_id: {prev}) that was auto-reset. If the user refers to "
+        f"(session_id: {prev}) that was auto-reset.{root_hint} If the user refers to "
         f"earlier work here, or the request depends on this {where}'s history, "
-        f"use the session_search tool to recall that prior session before "
+        f"use the session_search tool to recall the immediate prior session "
+        f"and, when the request points back to the original topic, the original "
+        f"lineage session before "
         f"acting — do not assume an unrelated recent session is the right "
         f"context.]"
     )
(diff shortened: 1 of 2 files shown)


Enough1122:

Improves continuity hints for multi-reset lineages without extra DB schema.

- **Root walk (`gateway/session.py:36+`):** `_continuity_root_session_id` walks `parent_session_id` chain with `max_depth=256`, cycle set, routing-key boundary check (`row_key != expected_session_key`), and malformed-row break. Preserves existing `parent_session_id` persistence, so no migration needed.
- **Note enrichment (`:9+`):** `build_channel_continuity_note` now appends `original session is session_id: {root}` only when `root != prev`, and wording clarifies to recall both immediate prior and original when request points to original topic. Avoids duplicate when lineage depth=1.
- **Persistence (`:80+`):** `continuity_root_session_id` computed once at candidate creation (`_get_or_create_session_impl`) from `prev_session_id`, stored in `metadata`. Cheap, single walk per reset.

Non-blocking:
- Walk uses `self._db.get_session` per hop; with `max_depth=256` worst-case 256 DB reads on deeply-reset channel. Acceptable given rare event, but consider early exit if `parent` equals `session_id` (self-loop) — already covered by `seen` set.

PR #87265, opened 2026-08-15
fix(telegram): prevent Windows long-poll socket reuse deadlock

JoaoMarcos44:
## Summary

Fixes #87057.

Telegram polling could become permanently silent on Windows after a transient network failure. The gateway process, cron ticker, and heartbeat remained alive, but inbound Telegram updates stopped because the reconnect path could reuse a stale `CLOSE-WAIT` socket.

This change fixes the failure at the transport/recovery boundary:

- the dedicated `getUpdates` pool never reuses a keep-alive connection;
- a timed-out polling drain uses the wall-clock deadline helper rather than waiting for cancellation-shielded PTB/httpcore cleanup;
- if PTB leaves `HTTPXRequest._client.is_closed == False` after the timeout, the polling client is replaced before the next polling generation starts;
- a stale-client cleanup that also absorbs cancellation is bounded and abandoned safely, preventing one live cleanup task from accumulating per reconnect;
- the general Bot API pool remains reusable and is not interrupted, preserving concurrent sends and edits.

Telegram long-poll recovery

> The infographic is hosted externally because this repository rejects committed PR-infographic binaries.

## Root cause

The existing recovery assumed that `HTTPXRequest.shutdown()` would either finish or leave the client marked closed. On the reported Windows failure, the underlying long-poll socket remained stuck in `CLOSE-WAIT`; `updater.stop()` and the polling request drain could exceed their await boundary while the HTTPX client still reported itself open.

PTB's `HTTPXRequest.initialize()` only builds a new client when the current client is closed. Consequently, the next `start_polling()` could inherit the same dead connection. Since the process itself was healthy, PID-based supervision could not detect the silent ingress failure.

A second contributing condition was connection reuse in the long-poll pool. `keepalive_expiry` only applies to idle connections, while a long poll remains active. A Telegram-side close could therefore be observed on the next poll before the pool had a chance to expire the socket.

### Dedicated polling transport

`getUpdates` now receives `httpx.Limits` with `max_keepalive_connections=0` on direct, proxy, and fallback-IP branches. The general request pool keeps the existing tuned reusable limits.

For the fallback-IP branch, the limits are passed to `TelegramFallbackTransport`, which owns the inner `AsyncHTTPTransport` instances. This avoids the known httpx behavior where client-level limits are ignored when a custom transport is supplied.

### Stale-client replacement

`_drain_polling_connections()` now:

1. bounds shutdown with `_await_with_thread_deadline()`;
2. detects a shutdown that did not complete;
[...]

PR #87111, opened 2026-08-15
fix(telegram): recover Windows CLOSE-WAIT getUpdates deadlock

HexLab98:
## Summary
- After a Telegram network error on Windows, `updater.stop()` can time out on a stale long-poll socket. PTB's `HTTPXRequest.initialize()` only rebuilds the httpx client when `is_closed` is true, so a hung `aclose()` left `start_polling()` on the same CLOSE-WAIT connection. The gateway process stayed alive (PID, cron, heartbeat) while inbound `getUpdates` blocked indefinitely ([#87057](https://github.com/NousResearch/hermes-agent/issues/87057)).
- Rebuild the polling HTTP client when drain times out, watch instrumented getUpdates I/O in the heartbeat (independent of `get_me()` / `pending_update_count`), and enable TCP keepalive on the fallback transport used for `api.telegram.org` IPs such as `149.154.166.110`.

## Test plan
- [x] `scripts/run_tests.sh tests/gateway/test_telegram_network_reconnect.py tests/gateway/test_telegram_network.py tests/gateway/test_telegram_polling_progress.py tests/gateway/test_telegram_closewait_limits_31599.py tests/test_telegram_polling_progress_ptb.py tests/gateway/test_telegram_init_deadline.py tests/gateway/test_telegram_pending_update_probe.py tests/gateway/test_telegram_conflict.py tests/gateway/test_telegram_start_polling_timeout.py tests/gateway/test_telegram_connect.py`
- [ ] Reproduce on Windows: force a Telegram `TimedOut` / `Bad Gateway`, confirm logs still show `updater.stop() timed out` then a client rebuild, and that new DMs arrive instead of a silent 48h hang.

PR #98641, opened 2026-08-30
fix(telegram): retain retired generations until cleanup completes

yazan1991:
## Summary

Telegram recovery can abandon PTB lifecycle work when `Updater.stop()`, application teardown, or related cleanup exceeds a bounded deadline. Today, Hermes can move on to a replacement adapter without retaining process-level ownership of the complete retired generation, leaving the old `Application`, `Updater`, `update_fetcher`, `Bot`, request clients, and abandoned lifecycle task without a final owner.

This PR keeps retired Telegram generations strongly owned until cleanup completes, caps retirement to one generation, serializes bounded cleanup, and drains any remaining retired generations during Gateway shutdown even if the Telegram adapter has already been removed from the active adapter map.

## What changes

- Add a loop-local, token-keyed retirement registry for abandoned Telegram generations.
- Retain the exact lifecycle task together with the retired `Application`, `Updater`, `Bot`, and request clients.
- Serialize and retry cleanup without overlapping PTB close steps.
- Bound retired-generation capacity to prevent linear resource growth across repeated reconnect failures.
- Preserve stale-callback isolation between old and current generations.
- Add an adapter-independent Gateway shutdown drain so ownership is finalized after adapter removal as well.
- Keep the existing fresh-generation replacement behavior. This PR does not introduce private `_client`/`_build_client` swapping.

## Why this is needed

Existing Telegram reconnect hardening covers bounded stop/drain behavior, polling liveness, fatal handoff, and transport/client recovery, but it does not provide a final process-level owner for lifecycle work that has already been abandoned by a deadline.

Without that ownership boundary, repeated `stop timeout -> abandon -> fatal recovery -> replacement` cycles can accumulate retired PTB generations even when stale callbacks are correctly fenced from mutating current state.

## Validation

Focused upstream tests on the rebased branch:

- 31 passed
- 1 pre-existing general-pool hang test deselected
- Ruff passed
- `compileall` passed

The focused suite covers:

- abandoned `Updater.stop()` ownership
- retired `Application` / `Updater` / `update_fetcher` lifetime
- ordered and idempotent cleanup
- one-generation capacity fencing
- stale callback isolation
- replacement without stale-generation reuse
- repeated failure/replacement resource plateau
- global shutdown drain
- shutdown drain after adapter removal
- no destroyed-pending-task warnings
- no unclosed client/transport warnings

Additional certification outside the committed CI test count exercised 100 replacement/failure cycles across PTB 22.6 and 22.8 and returned retired resources to zero after shutdown.

## Scope

Runtime changes are limited to:

- `plugins/platforms/telegram/adapter.py`
- `gateway/run.py`

Tests are limited to:

- `tests/gateway/test_telegram_network_reconnect.py`
- `tests/gateway/test_telegram_retired_generation_v2.py`

This is intentionally separate from the Telegram reconnect config-cache fix in #98235, which addresses synchronous config parsing during adapter rebuild rather than retired-generation ownership.

Related existing work such as #87265 and #87111 focuses on transport/client recovery and does not retain the complete abandoned PTB generation through final cleanup.

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -2810,6 +2810,17 @@ from gateway.whatsapp_identity import (
 logger = logging.getLogger(__name__)
 
 
+async def _drain_global_telegram_retired_generations() -> bool:
+    """Drain Telegram retirement ownership even after adapter removal."""
+    try:
+        from plugins.platforms.telegram.adapter import (
+            drain_telegram_retired_generations,
+        )
+    except ImportError:
+        return True
+    return await drain_telegram_retired_generations()
+
+
 _OWN_POLICY_OPEN_ENV = {
     Platform.WECOM: ("WECOM_DM_POLICY", "WECOM_GROUP_POLICY", "WECOM_ALLOW_ALL_USERS"),
     Platform.WEIXIN: ("WEIXIN_DM_POLICY", "WEIXIN_GROUP_POLICY", "WEIXIN_ALLOW_ALL_USERS"),
@@ -15818,6 +15829,19 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                 "Shutdown phase: all adapters disconnected at +%.2fs",
                 _phase_elapsed(),
             )
+            try:
+                telegram_retired_drained = (
+                    await _drain_global_telegram_retired_generations()
+                )
+                if not telegram_retired_drained:
+                    logger.error(
+                        "Gateway shutdown left bounded Telegram retired-generation "
+                        "ownership pending"
+                    )
+            except Exception:
+                logger.exception(
+                    "Gateway global Telegram retired-generation drain failed"
+                )
 
             for _task in list(self._background_tasks):
                 if _task is self._stop_task:
--- a/plugins/platforms/telegram/adapter.py
+++ b/plugins/platforms/telegram/adapter.py
@@ -15,7 +15,6 @@ import logging
 import os
 import html as _html
 import re
-import threading
 import time
 from contextvars import ContextVar
 from datetime import datetime, timezone
@@ -24,6 +23,14 @@ from typing import Any, Awaitable, Callable, Dict, List, Optional, Set
 logger = logging.getLogger(__name__)
 
 from agent.deadline import run_bounded_async
+from plugins.platforms.telegram.retirement import (
+    CAPACITY_WAIT_TIMEOUT as _RETIRED_CAPACITY_WAIT_TIMEOUT,
+    RetiredTelegramGeneration as _RetiredTelegramGeneration,
+    RetirementCapacityError,
+    TelegramRetirementRegistry as _TelegramRetirementRegistry,
+    drain_retired_generations as drain_telegram_retired_generations,
+    get_retirement_registry as _get_telegram_retirement_registry,
+)
 
 
 def _redact_telegram_error_text(error: object) -> str:
@@ -102,43 +109,6 @@ async def _first_completed(*futures: "asyncio.Future") -> None:
     """
     await asyncio.wait(set(futures), return_when=asyncio.FIRST_COMPLETED)
 
-
-async def _shutdown_abandoned_app(app) -> None:
-    """Release a half-built PTB app's httpx transports after init was abandoned.
-
-    ``Application.shutdown()`` / ``Bot.shutdown()`` are gated on the app's
-    ``_initialized`` / ``_requests_initialized`` flags, which a wedged
-    ``initialize()`` (the case this whole path exists for) may never have set —
-    so calling only ``app.shutdown()`` no-ops and leaks the connection pool it
-    was meant to close.  ``HTTPXRequest`` builds its ``httpx.AsyncClient``
-    eagerly in its constructor and its ``shutdown()`` gates only on
-    ``client.is_closed``, so closing the request transports directly releases
-    the pool regardless of PTB init state.  We try the clean path first, then
-    fall back to the transports.  All best-effort and swallowed.
-    """
-    if app is None:
-        return
-    try:
-        await app.shutdown()
-    except Exception:
-        logger.debug("Abandoned Telegram app.shutdown() failed", exc_info=True)
-    # Directly close the underlying request transports (bypasses PTB's
-    # init-gated shutdown so the eagerly-built httpx pool is released even when
-    # the abandoned initialize() never flipped _initialized).
-    bot = getattr(app, "bot", None)
-    requests = getattr(bot, "_request", None) if bot is not None else None
-    if not requests:
-        return
-    for request in requests:
-        shutdown = getattr(request, "shutdown", None)
-        if shutdown is None:
-            continue
-        try:
-            result = shutdown()
-            if asyncio.iscoroutine(result) or asyncio.isfuture(result):
-                await result
-        except Exception:
-            logger.debug("Abandoned Telegram request shutdown failed", exc_info=True)
 
 try:
     from telegram import Update, Bot, Message, InlineKeyboardButton, InlineKeyboardMarkup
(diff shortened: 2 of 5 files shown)


andrexibiza (commented):
Reviewed exact head `7891fea23b71ae6e02324ffefbb6f157504a214a` against current `main@5cc1369fa298021f8c740de154ff8c37c30bdcc8` (one submitted commit, directly parented by current main). The central direction is good: timeout is no longer treated as disposal, the exact abandoned lifecycle task travels with the retired PTB object graph, cleanup steps are serialized/idempotent, and shutdown gets an adapter-independent final drain. That is a much stronger ownership model than best-effort detached cleanup.

I found two correctness blockers in the ownership state machine, plus landing/merge-order gates.

### P1 — the one-slot invariant is bypassed *inside* the connect retry ladder, so a second abandoned generation can become unowned

The new public-connect gate is correct for a registry that is already full before `connect()` starts, and `test_real_connect_gate_refuses_new_application_at_capacity` proves that case. The other side of the shape is the retry ladder itself:

1. Generation A is built and `Application.initialize()` exceeds `_init_timeout`.
2. `_await_owned_lifecycle()` correctly retires A and tracks A's still-running initialize task in the sole registry slot.
3. In the `rebuild_app` branch, `registry._find(old_app)` now finds A, so the code deliberately skips `wait_for_capacity()` and immediately does `self._app = builder.build()` for generation B while A still owns the only retirement slot.
4. If B's initialize/start/stop/delete-webhook lifecycle await is then abandoned before A finishes, `_await_owned_lifecycle()` calls `_retire_app(B, start=False)`. `retire()` returns `None` because `len(entries) >= _MAX_RETIRED_GENERATIONS`, and the B task is therefore never attached to any owner record before the exception is re-raised.

That recreates the exact defect class this PR is intended to close: an abandoned PTB generation with no final process owner. There is a second symptom of the same state: `_handoff_polling_fatal_error()` retries `wait_for_capacity(5s)` in an unbounded `while entry is None` loop, so a fatal notification can remain pinned behind an older retirement indefinitely. That is particularly important in the lineage of #80598, where the production failure was a fatal handler that never reached the reconnect queue.

Please make the invariant structural: either **do not construct a replacement `Application` until the token's retired generation has actually released capacity**, or change the ownership model so the currently-live generation can always transfer into bounded retirement without depending on a slot occupied by its predecessor. Whichever shape you choose, the fatal handoff still needs one total deadline / typed failure path rather than repeated 5-second waits with no global bound.

The regression I want here is adversarial and exact: hold A's abandoned initialize/cleanup task open, enter the in-connect rebuild to B, abandon B as well, and prove that (a) no lifecycle task/object graph is unowned, (b) no generation C is created, (c) the fatal/reconnect handoff remains bounded, and (d) releasing A/B drains to zero. The current capacity test does not exercise this reachable two-consecutive-abandonment path.

### P1 — the global registry is not actually weak and token churn is retained forever

`_RETIREMENT_REGISTRIES` is a `WeakKeyDictionary` keyed by the event loop, but every `_TelegramRetirementRegistry` stored in the value dict keeps `self.loop = loop` strongly. The reachability is therefore:

`WeakKeyDictionary -> per_loop dict -> registry -> loop (the weak key)`.

A weak-key container cannot evict a key while one of its own values strongly retains that key. Separately, when an entry reaches `CLEANED`, it is removed from `registry.entries`, but the token-hash → registry mapping is never removed from `per_loop`. A long-lived gateway therefore accumulates one registry per token ever seen, and a finished event loop can remain reachable through its registry values.

`test_global_shutdown_drain_repeats_without_retention` currently creates five distinct token registries and checks only that each registry's `entries` list/resource graph becomes empty; it never checks the global per-loop map cardinality or that the loop itself becomes collectible. Please make empty registries removable (and avoid a strong loop back-reference from a weak-key value), then add both a many-token churn assertion and a loop weakref/GC assertion.

### Ownership / merge-order gate

This PR adds the new retirement state machine directly to `plugins/platforms/telegram/adapter.py` (already thousands of lines; this patch reaches past line 5,600) and adds Telegram-specific shutdown ownership to `gateway/run.py` (patch reaches past line 15,800). These are already collision-heavy owners. I would extract the retirement registry/state machine into a bounded Telegram lifecycle module and keep only the narrow adapter/gateway seams in the godfiles before landing.

The live interlock graph matters here:

- #98235 is **complementary**, not duplicate: it fixes synchronous config parsing during adapter rebuild and should remain independent.
- #87265 (JoaoMarcos44) and #87111 (HexLab98) are **competing/adjacent realizations at the same polling-drain seam**. This PR's whole-generation retirement can supersede their private-client replacement *strategy*, but not their distinct transport/liveness invariants (notably no-reuse/keepalive/getUpdates-liveness work). Preserve those pieces and their credit if this becomes the convergence object.
- #59336 (rayjun) is **adjacent webhook-start lifetime work** and currently builds its abandonment cleanup around `_shutdown_abandoned_app`, which this PR deletes. If this lands first, that webhook deadline needs recomposition onto the retirement owner rather than being silently lost.
- #97446 is **complementary durable-ingress ownership** but directly collides in `plugins/platforms/telegram/adapter.py`; app retirement must not break its queue/claim handoff semantics.
- #67455 is **complementary multi-account Telegram architecture** and collides in both runtime files. Because this registry is token-keyed, that PR's account identity/lifecycle work must be recomposed explicitly, not merged by whichever branch happens to land last.

Please preserve the production provenance in #67498/#80598 when closing or superseding older recovery threads; those reports established the idle-loop/orphaned-lifecycle and fatal-handler-wedge failure shapes this work is trying to close.

### Exact-head acceptance / attribution

There is no hosted acceptance receipt yet. Exact-head CI `33318201748`, Docker `33318201303`, and Nix `33318201312` all concluded `action_required`; the CI run currently has **0 jobs**. The reported `31 passed` and 100-cycle PTB 22.6/22.8 soak are useful construction evidence, but they do not substitute for exact-head hosted acceptance.

There is also a deterministic contributor-attribution failure waiting behind that workflow gate: the sole submitted commit is authored and committed as `Codex <codex@localhost>`. Current `main`'s `Contributor Attribution Check` rejects any new non-exempt email without a `contributors/emails/<email>` mapping or legacy mapping, and current main contains no `codex@localhost` mapping. Please bind the commit to the actual contributing human identity, or add the explicit mapping if that machine-local identity is intentionally meant to resolve to a human contributor, then run the repository attribution audit and get fresh exact-head CI.

The core idea is strong and worth converging: **abandonment must transfer ownership, not erase it**. The remaining work is making that true for the second failure in a row, for the lifetime of the registry itself, and across the adjacent Telegram recovery objects rather than only for the first retired generation.

yazan1991:
Thanks for the detailed review. I pushed `934469b83e4f1864002ea28e1aaa393327d03e97` rebased on current `main@4f22543509d1b91dc45bcb369447126c5eb14fb7` and addressed both P1 blockers plus the landing items you called out.

### P1: second consecutive abandonment / bounded fatal handoff

The capacity invariant is now structural: `connect()` does not construct the next PTB `Application` while the token's existing retired generation still occupies the sole slot. The in-connect rebuild ladder uses the same gate, so a second abandoned generation cannot be created over an occupied predecessor.

The fatal handoff no longer loops on repeated 5s capacity waits. It has one total capacity budget and raises `RetirementCapacityError` when that budget expires.

Added an adversarial regression that holds generation A's abandoned lifecycle work open, enters the rebuild path, verifies B is not constructed until A drains, then abandons B and proves B remains owned and no generation C is created. Releasing the held work drains retirement to zero. A separate test exercises the real fatal/reconnect path with a short capacity budget and verifies the typed bounded failure.

### P1: weak registry / token churn

The retirement registry now stores only `weakref.ref(loop)`, so the value no longer strongly retains the `WeakKeyDictionary` key. Empty token registries are removed with identity + emptiness checks, and empty per-loop maps are removed as well.

Added regressions for 32-token churn returning registry cardinality to zero after cleanup and for a disposable closed event loop becoming collectible after GC.

### Structure / attribution

The retirement state machine was extracted to `plugins/platforms/telegram/retirement.py`; `adapter.py` now keeps the Telegram integration seams and `gateway/run.py` keeps only the shutdown drain seam.

The commit is now attributed to `yazan1991` rather than `Codex <codex@localhost>`, and the repository attribution check passes locally.

Focused validation on the updated head: 50 passed, 1 pre-existing general-pool hang test deselected; reconnect lane 19 passed/1 deselected; adjacent connect/start/pending-update/PTB/polling-progress lanes 24 passed/1 deselected; Ruff and compileall pass. The existing transport/liveness, webhook, durable-ingress, and multi-account adjacent PR invariants were rechecked and no concrete interlock regression was found.

Hosted workflows on this exact head are still reporting `action_required` with no acceptance jobs, so I am not treating local validation as hosted CI acceptance.

andrexibiza (commented):
Re-review of exact head `934469b83e4f1864002ea28e1aaa393327d03e97` against exact base `4f22543509d1b91dc45bcb369447126c5eb14fb7`.

The previous exact-capacity bypass, repeated fatal-capacity waits, weak loop/token registry retention, state-machine extraction, and commit-author provenance defects are materially addressed. This head is still not ready to land.

## [P1] The retry ladder reattaches the descendants that retirement just closed

`connect()` constructs one `ApplicationBuilder`, one general `HTTPXRequest`, one polling `HTTPXRequest`, and—in the fallback-IP branch—two concrete `TelegramFallbackTransport` instances, then builds generation A from them. After A is fully retired, the retry path creates B with the same `builder.build()` call rather than constructing a new PTB graph:

- [builder/request/transport construction](https://github.com/NousResearch/hermes-agent/blob/934469b83e4f1864002ea28e1aaa393327d03e97/plugins/platforms/telegram/adapter.py#L4568-L4778)
- [builder reuse after retirement](https://github.com/NousResearch/hermes-agent/blob/934469b83e4f1864002ea28e1aaa393327d03e97/plugins/platforms/telegram/adapter.py#L4873-L4897)

That is not a fresh generation under PTB 22.8. `ApplicationBuilder._build_request()` returns the exact custom request object stored on the builder, and every subsequent `build()` passes those same objects into the new `ExtBot`; the builder also reuses its stored update queue and job/update-processor objects: [PTB 22.8 builder source](https://github.com/python-telegram-bot/python-telegram-bot/blob/ac42b53/src/telegram/ext/_applicationbuilder.py#L225-L337). `HTTPXRequest.initialize()` only rebuilds the `AsyncClient` from the same `_client_kwargs`, which retain the same custom transport object: [PTB 22.8 HTTPXRequest source](https://github.com/python-telegram-bot/python-telegram-bot/blob/ac42b53/src/telegram/request/_httpxrequest.py#L176-L229).

This PR’s retirement owner then explicitly shuts those exact request objects: [retirement cleanup](https://github.com/NousResearch/hermes-agent/blob/934469b83e4f1864002ea28e1aaa393327d03e97/plugins/platforms/telegram/retirement.py#L245-L310). In the fallback path, `TelegramFallbackTransport.aclose()` permanently marks the primary closed and closes its pools: [transport close](https://github.com/NousResearch/hermes-agent/blob/4f22543509d1b91dc45bcb369447126c5eb14fb7/plugins/platforms/telegram/telegram_network.py#L198-L208). B therefore inherits A’s retired request/transport identities. A custom `base_url` routes directly through that closed primary; hostname-last recovery on an IPv6-only path also no longer has a live primary to fall back to. More fundamentally, the old and replacement applications are not disjoint ownership generations.

The new consecutive-abandonment test cannot detect this because it injects already-disjoint `_ConnectApp` objects behind a `MagicMock` builder and explicitly disables fallback transports: [test](https://github.com/NousResearch/hermes-agent/blob/934469b83e4f1864002ea28e1aaa393327d03e97/tests/gateway/test_telegram_retired_generation_v2.py#L273-L374).

Required before landing:

1. Build the complete PTB resource graph from a generation factory after retirement capacity is actually free: new builder, general request, polling request, fallback transports, update queue, and other builder-owned mutable runtime objects.
2. Add an identity-faithful or real-PTB regression proving A and B do not share either request, either fallback transport, or the update queue, and proving B initializes after A has been fully retired. Cover the fallback-IP/custom-base-URL path rather than disabling it.

## [P1] Capacity expiry drops the only Gateway fatal handoff

`_handoff_polling_fatal_error()` now has one bounded capacity wait, but on expiry it raises `RetirementCapacityError` before clearing the carrier and before `_notify_fatal_error()`:

- [fatal handoff](https://github.com/NousResearch/hermes-agent/blob/934469b83e4f1864002ea28e1aaa393327d03e97/plugins/platforms/telegram/adapter.py#L3830-L3868)
- [recovery task is only discarded on completion](https://github.com/NousResearch/hermes-agent/blob/934469b83e4f1864002ea28e1aaa393327d03e97/plugins/platforms/telegram/adapter.py#L2835-L2857)

At that point the adapter is already retryable-fatal. The detached `_polling_error_task` has no continuation that reacquires capacity or routes the exception to the Gateway, and `has_fatal_error` suppresses later recovery scheduling. The base notification is specifically the shielded path that completes adapter removal and reconnect queueing; bypassing it is the zombie-gateway failure mode documented in [the base adapter](https://github.com/NousResearch/hermes-agent/blob/4f22543509d1b91dc45bcb369447126c5eb14fb7/gateway/platforms/base.py#L3630-L3668).

The regression currently codifies the dropped handoff with `assert_not_awaited()` and then performs a manual retirement that production never schedules: [test](https://github.com/NousResearch/hermes-agent/blob/934469b83e4f1864002ea28e1aaa393327d03e97/tests/gateway/test_telegram_retired_generation_v2.py#L376-L425).

Keep the caller-visible wait bounded, but transfer responsibility to a strongly rooted, exactly-once continuation. The current application must remain owned until the slot frees; then it must enter retirement and invoke the Gateway fatal handler exactly once. The regression must hold the old slot past the first deadline, prove there is no repeated synchronous wait or unhandled task exception, release the slot, and prove the fatal handler and cleanup both complete without manual intervention. Calling the fatal handler before ownership is secured is not sufficient because Gateway teardown may immediately pop/cancel the adapter while the retirement slot is still occupied.

## Proof and landing gates

The exact-head CI, Docker, and Nix workflows are all `action_required`, so this object has no hosted exact-head acceptance yet:

- [CI](https://github.com/NousResearch/hermes-agent/actions/runs/33323127787)
- [Docker](https://github.com/NousResearch/hermes-agent/actions/runs/33323127491)
- [Nix](https://github.com/NousResearch/hermes-agent/actions/runs/33323127528)

The PR body is also stale relative to this head: production scope omits `plugins/platforms/telegram/retirement.py`, committed validation still says 31 tests while the head comment reports 50, and the related-work section does not define composition/merge order. Before landing, update the body and add an explicit intersection table for at least #98235, #87265, #87111, #59336, #97446, and #67455: overlapping seams, ownership, merge order, rebase/revalidation plan, and prior-work credit. Every commit must then acquire exact-object hosted proof.

PR #91371, opened 2026-08-21
feat(desktop): add Turkish (tr) locale

yunusemrec80:
## Summary

Adds Turkish (`tr`) as a desktop UI language.

- **`tr.ts`**: ~2,500 translated strings, built as a partial locale via `defineLocale` so any untranslated key falls back to English.
- Registers `'tr'` in the `Locale` union (`types.ts`).
- Adds Turkish to `LOCALE_OPTIONS` and `LOCALE_ALIASES` (`languages.ts`).
- Imports and registers `tr` in `TRANSLATIONS` (`catalog.ts`).

The CLI Turkish locale (`locales/tr.yaml`) already exists upstream, so this completes Turkish support on the desktop surface.

## Notes

- Partial coverage: roughly 2,500 of ~2,900 keys are translated; the remainder fall back to English. More strings can be added incrementally in follow-ups.
- Function-valued (parameterized) translation keys keep their parameter names and `${...}` placeholders, with only the surrounding text translated.

PR #87305, opened 2026-08-15
feat(desktop): add Turkish (tr) localization

canlancan1339-wq:
## Summary

Adds a complete **Turkish (`tr`)** localization for the Hermes desktop app UI.

## What's included

- **`apps/desktop/src/i18n/tr.ts`** (new) — full Turkish translations for every string in the desktop catalog (uses the `defineLocale` partial-override pattern like `ja.ts` and `ar.ts`, so any missing key falls back to English).
- **`types.ts`** — add `'tr'` to the `Locale` union.
- **`languages.ts`** — register `Türkçe` in the locale picker (`LOCALE_OPTIONS`) plus normalization aliases (`tr`, `tr-tr`, `tr_tr`, `turkish`, `türkçe`).
- **`catalog.ts`** — wire `tr` into the `TRANSLATIONS` registry.

Turkish is LTR, so no `RTL_LOCALES` change is needed.

## Verification

- `tsc --build apps/desktop/tsconfig.json` passes (exit 0).
- Key-diff against `en.ts` confirms zero untranslated keys remaining.

Closes a gap where the desktop app supported `en/zh/zh-hant/ja/ar` but not Turkish.

PR #57218, opened 2026-07-02
feat(i18n): add Turkish (tr) locale for Hermes Desktop

rasitnarcicek:
This PR adds Turkish (tr) locale support for Hermes Desktop.

**Changes:**
- Added `apps/desktop/src/i18n/tr.ts` with Turkish translations (defineLocale, fallback to English)
- Updated `types.ts`: added `'tr'` to Locale type
- Updated `languages.ts`: added Turkish to LOCALE_OPTIONS and LOCALE_ALIASES
- Updated `catalog.ts`: registered tr in TRANSLATIONS
- Updated `locales/tr.yaml` from upstream

**Status:**
- Common UI strings translated (apply, save, cancel, etc.)
- Language settings translated
- Missing strings fall back to English via defineLocale()
- ~50% translated initially, community can complete gradually

PR #46329, opened 2026-06-14
feat(i18n): add Turkish (tr) locale for desktop app

sprmn24:
## What does this PR do?
Adds full Turkish (`tr`) locale support for Hermes Desktop. Turkish is currently missing from the desktop app's language options, even though it's already supported in the web dashboard (`web/src/i18n/tr.ts`, PR #20474).

## Related Issue
Fixes #

## Type of Change
- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 🔒 Security fix
- [ ] 📝 Documentation update
- [ ] ✅ Tests (adding or improving test coverage)
- [ ] ♻️ Refactor (no behavior change)
- [ ] 🎯 New skill (bundled or hub)

## Changes Made
- `apps/desktop/src/i18n/tr.ts` (new): Full Turkish translation covering all 28 top-level sections of `en.ts` (common, boot, notifications, titlebar, keybinds, language, settings, skills, agents, commandCenter, messaging, profiles, cron, artifacts, sidebar, composer, statusStack, updates, install, onboarding, modelPicker, modelVisibility, shell, rightSidebar, preview, assistant, prompts, desktop, errors, ui), following the same `defineLocale`/`defineFieldCopy` pattern as `ja.ts`.
- `apps/desktop/src/i18n/types.ts`: Added `'tr'` to the `Locale` type.
- `apps/desktop/src/i18n/languages.ts`: Added Turkish entry to `LOCALE_OPTIONS` and `tr`/`tr-tr`/`tr_tr` to `LOCALE_ALIASES`.
- `apps/desktop/src/i18n/catalog.ts`: Imported and registered `tr` in the locale catalog.

## How to Test
1. Run `npx tsc --noEmit -p apps/desktop/tsconfig.json` — compiles with no new type errors.
2. In Hermes Desktop, open Settings → Language and select Türkçe.
3. Verify UI strings across the app (titlebar, sidebar, settings, onboarding, etc.) render in Turkish with no missing-key fallbacks to English.

### Code
- [x] I've read the [Contributing Guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md)
- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(scope):`, `feat(scope):`, etc.)
- [x] I searched for [existing PRs](https://github.com/NousResearch/hermes-agent/pulls) to make sure this isn't a duplicate
[...]

PR #98642, opened 2026-08-30
feat(desktop): add Turkish (tr) locale

statukocu:
Adds Turkish to the desktop app — `tr.ts`, the catalog/picker/alias wiring, and the Kanban and Bot Mode plugin bundles. `agent/i18n.py` and the web UI already listed `tr`, so this is the desktop side catching up.

A couple of things that aren't just translation:

- The new-project idea pills are content, not chrome, so `randomIdeaTemplates()` now takes a locale and picks from a Turkish pool. A test asserts every entry is actually translated, so adding a 19th English template without a Turkish one fails instead of silently shipping an English pill.
- Added `tr` to the three existing `BOTS_LOCALES` parity tests. The interpolator one is the reason — types can't catch a translation that drops its argument.

Terminology I'd like a second opinion on: the app uses both "Skill" and "Capabilities" and they mean different things, so they're "beceri" and "yetenek" respectively rather than collapsing into one word. "Artifacts" follows the "çıktı" that the Kanban bundle already used.

## Testing

`npm run typecheck`, `npm run lint` and `vitest run --project ui` are clean on Windows 11.

I have not run the desktop app against this yet, so the strings are unverified in situ — happy to do that pass if you'd rather it land verified. Nothing here is platform-specific, and it's additive: no existing locale changes behaviour.


alt-glitch:

Related: this is another complete registered Turkish Desktop locale alongside #46329, #57218, #87305, and #91371. It has a concrete broader delta (plugin-bundle parity and localized project ideas), so this is a competing implementation rather than a duplicate; maintainer selection is needed.

sprmn24 (commented):
Native Turkish speaker here (I authored the earlier #46329). Reviewed `tr.ts` at `fc2dc4d` — overall quality is good: register is consistent (short imperatives on buttons, formal "siz" in sentences), and the contract changes that tripped up older Turkish PRs (`remoteSignInHint`, `branchTitle`) are already parameterized correctly.

**Terminology — agree with your calls**
- Skill → "beceri", Capabilities → "yetenek": 👍 keeping them distinct is right.

**Should fix**
1. `tr.ts:1009` (`removeConfirmDesc`) — "Örneğin kendisi etkilenmez" reads as *"for example, itself is not affected"*, since "örneğin" is the everyday word for "for example". Suggest:
   `…bu uygulamadan kaldırılacak. Sunucu örneğinin kendisi etkilenmez — …`
2. Artifacts uses two terms: "Yapıtlar" (`nav.artifacts`, artifacts page `:2372-2382`) vs "Çıktılar" (command-center tile `:1772`). "Çıktı" is also used for tool/installer *output* in the same file (`:519`, `:1302`, `:2972`), so I'd standardize on **"Yapıt"** everywhere to avoid the collision (or on "Çıktı" everywhere if you prefer Kanban parity — either way, one term).
3. `:1770` — Capabilities tile detail is stale vs `en.ts:1975`: missing "and plugins" → `'Beceriler, araçlar, MCP sunucuları ve eklentiler'`.

**Coverage note (non-blocking)**
These top-level sections currently fall back to English: `connectors`, `freeTier`, `guidedGreeting`, `handoffTour`, `sessionImport` (onboarding-facing, so fairly visible). Overall coverage is in line with `ja`/`ru`, so fine to land as-is and follow up if preferred.

**Nit**
"Oturum" covers both *sign-in* and *agent session*, which yields strings like "Oturumu kapat ve oturum aç". Understandable, but "Giriş yap / Çıkış yap" for auth would disambiguate. Up to you.

Happy to re-review after changes.

PR #96887, opened 2026-08-28
test(desktop): make locale-sensitive UI suites deterministic

Agi-Asi:
## What

Building/testing on a host with a non-Latin numeric locale (`fa_IR.UTF-8`) failed UI tests: the suites assert en-US-shaped output (`'1,234'`) while the formatters under test deliberately render in the **host** locale (`'۱٬۲۳۴'` under Persian). The production behaviour is correct — a dashboard formats numbers for its user's locale — the tests were environment-dependent.

Pin `Number.prototype.toLocaleString` to `'en-US'` for the two affected suites (`use-prompt-actions` utils, `fallback-model`) and restore it afterwards, so the assertions are deterministic on any host **without changing user-facing formatting**.

## Reproduction

```
LANG=fa_IR.UTF-8 node -e "console.log((1234567).toLocaleString())"
→ ۱٬۲۳۴٬۵۶۷
```

`LANG=fa_IR.UTF-8 npx vitest run --project ui` failed 3 tests across 2 files on current main (the issue's 4th failure no longer reproduces on main — upstream drift since v0.20.5); all three are covered by this pin. Same suites are green on an unmodified host.

## How to test

```
cd apps/desktop
LANG=fa_IR.UTF-8 npx vitest run src/app/session/hooks/use-prompt-actions/utils.test.ts src/components/assistant-ui/tool/fallback-model.test.ts --project ui   # 89 passed
npx vitest run src/app/session/hooks/use-prompt-actions/utils.test.ts src/components/assistant-ui/tool/fallback-model.test.ts --project ui                     # 89 passed
```

`eslint` clean (both files).

## Platforms

Linux (vitest). Test-only change — zero production code touched.

Fixes #96382

PR #98643, opened 2026-08-30
test(desktop): pin the UI suite's default locale to en-US

statukocu:
Two UI tests fail on a clean checkout if your machine isn't set to an en-US locale:

```
FAIL  src/app/session/hooks/use-prompt-actions/utils.test.ts
  expected 'Usage: 12 calls · 1.234.567 in / …'
        to be 'Usage: 12 calls · 1,234,567 in / …'

FAIL  src/components/assistant-ui/tool/fallback-model.test.ts
  expected '…' to contain '5,000 more characters truncated'
```

The app formats numbers against the host locale on purpose (`Intl` / `toLocaleString` with no locale argument — same as `src/lib/time.ts`), so this is the tests assuming a locale, not a product bug. CI's runners are en-US so it never showed up there.

`LANG` / `LC_ALL` don't help portably — ICU takes the locale from the OS on Windows and ignores them, which I confirmed before reaching for this:

```
$ LC_ALL=en_US.UTF-8 node -e "console.log((1234567).toLocaleString())"
1.234.567
```

So the default gets pinned in `vitest.setup.ts` instead. Covers the prototype `toLocale*String` methods and the `Intl` constructors, since `new Intl.NumberFormat(undefined, …)` resolves the same way. The `Intl` side uses a `Proxy` so `instanceof` and `supportedLocalesOf` still work.

## Testing

On a tr-TR Windows 11 box both tests pass now. The rest of `vitest run --project ui`
is unaffected — 6756 green. I did have to raise `testTimeout` locally to get a clean
full run, but that's cold-start slowness on this machine and unrelated to this change,
so it's not in the diff.

### DIFF
--- a/apps/desktop/vitest.setup.ts
+++ b/apps/desktop/vitest.setup.ts
@@ -38,3 +38,38 @@ if (typeof (globalThis as any).localStorage === 'undefined') {
 // CPU contention in CI. Success still resolves the instant the node appears;
 // the wider deadline only absorbs a starved runner, killing timing flakes.
 configure({ asyncUtilTimeout: 5000 })
+
+// Number and date formatting in the app deliberately follows the host locale
+// (`Intl.*` and `toLocaleString()` called with no locale — see src/lib/time.ts).
+// Tests assert en-US output, which holds silently on CI's en-US runners and
+// breaks for any contributor whose machine is set to another locale: tr-TR
+// renders 1.234.567, de-DE renders 1.234.567, fr-FR renders 1 234 567. An env
+// var cannot fix this portably — ICU reads the locale from the OS on Windows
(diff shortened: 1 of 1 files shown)


alt-glitch:

Related: #96887 already repairs the same two locale-sensitive tests by pinning and restoring the locale within each file. This PR uses a global Vitest setup approach; both are valid competing isolation strategies.

OutThisLife:
Superseded by https://github.com/NousResearch/hermes-agent/pull/120267. The replacement makes the affected assertions locale-aware without globally replacing Intl or prototype formatters, so non-English host behavior remains exercised. Thanks @statukocu for documenting the Windows locale behavior and the alternative test-isolation approach.

PR #98644, opened 2026-08-30
perf(desktop): measure compositing, and stop a sidebar full of working agents from growing the layer tree

Dolverin:
## Behavior
- Bound sidebar working-ring animation using the union of working and stalled session IDs; overlap does not double-count rows.
- Preserve visible working rings and accessibility treatment when over-cap rings become static.
- Remove scoped CDP listeners on success and failure, with idempotent unsubscribe behavior even for duplicate callback registration.
- Add compositing measurement helpers and fixture discovery without asserting measured hardware performance.

## Validation
- Focused UI/store/row tests: **39 passed**.
- Compositing helper fixtures: **10 passed**.
- Typecheck, changed-file ESLint and staged diff check passed; exact ten-file candidate received independent final review.
- The broad UI suite is **not claimed green**: bounded candidate and clean-base runs timed out, and isolated replay reproduced the same four failures with 114 passes on both. These are outside this delta.
- **No live Electron/GPU benchmark was completed. No measured performance improvement is claimed.** Fixture tests demonstrate logic, not real compositor performance.

## Publication
Updated fork revision: `730c9a493bbb0a8aab4d59558ce28ed3a93bd1f5`, one reviewed commit on upstream base `91adf584a4783d600e8f358a4f32ad95e7e0cdde`. Hosted CI must be evaluated for this exact revision; previous checks do not validate it. No installation or runtime change is included.

### DIFF
--- a/apps/desktop/scripts/perf/lib/cdp.mjs
+++ b/apps/desktop/scripts/perf/lib/cdp.mjs
@@ -124,6 +124,32 @@ export class CDP {
     }
 
     this.listeners.get(method).push(handler)
+    let subscribed = true
+
+    return () => {
+      if (!subscribed) {
+        return
+      }
+
+      subscribed = false
+      const handlers = this.listeners.get(method)
+
+      if (!handlers) {
+        return
+      }
+
+      const index = handlers.indexOf(handler)
+
+      if (index === -1) {
+        return
+      }
+
+      handlers.splice(index, 1)
+
+      if (handlers.length === 0) {
+        this.listeners.delete(method)
+      }
+    }
   }
 
   /** Evaluate an expression in the page and return its value (awaits promises). */
--- a/apps/desktop/scripts/perf/lib/compositing.mjs
+++ b/apps/desktop/scripts/perf/lib/compositing.mjs
@@ -1,0 +1,206 @@
+// Collect rendering timeline phases and layer-tree observations alongside the
+// harness's JavaScript, commit-count and frame-pacing metrics. These helpers
+// report what a scenario emits; their synthetic tests are not hardware benchmarks.
+//
+// Usage mirrors withCpuProfile:
+//
+//   const { result, compositing } = await withCompositingTrace(cdp, () => driveScenario())
+//   const layers = await layerSnapshot(cdp)
+
+import { sleep } from './cdp.mjs'
+
+/** Timeline events worth attributing. Names are stable devtools.timeline phases. */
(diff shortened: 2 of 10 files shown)


Enough1122:

Strong perf instrumentation + fix for compositing layer explosion. Measured improvement (Layerize 3.2%→19.9% at busy=16, layers 29+4·N) justifies the cap.

- **Compositing harness (`apps/desktop/scripts/perf/lib/compositing.mjs:45+`):** `aggregateTraceEvents` correctly filters `ph=X` and aggregates `ms/pct/n/max_ms`, with wallUs floor to avoid div/0. `withCompositingTrace` properly closes trace in `finally` and awaits `tracingComplete` to avoid poisoning later scenarios. Limiting to `devtools.timeline` avoids 173k UpdateLayer noise — pragmatic.
- **Layer snapshot (`:143+`):** `layerSnapshot` settles then samples `reasonSample=60` compositing reasons, tolerates vanishing layers. `watchLayerTreeRate` separates load-window churn from settle-window — prior bug fix noted in comments.
- **Cap (`apps/desktop/src/store/session-dot-state.ts:74+`):** `MAX_ANIMATED_RUNNING_ARCS=4`, `computed([$working,$stalled])` => `<=4`. All rings go still together, avoiding order-dependent flicker. `session-row.tsx:470+` renders `arc-still` when `!arcsAnimated`.
- **CSS correctness (`apps/desktop/src/styles.css:632+`):** `.arc-still::before {animation:none; will-change:auto}` — comment explains `will-change` is load-bearing; removing only animation left layers unchanged (measured 86 layers at 16 running). Also handles `data-renderer-animations-paused`.

Non-blocking:
- `vitest.config.ts:649` broadens include to `scripts/**/*.test.{ts,mjs}` — verify `run-short-session-hang-repro` exclusion still correct under new glob.

jerrygooch:
Checked the sidebar compositing fix. Logic is sound and well-tested.

When more than four session rings are animating (`working + stalled > 4`), all rings go still together via `arc-still`. That cap avoids piling up animated layers when you have a full house of agents, and doing it together prevents a flickery half-animated state.

The per-row subscription via `useStore` means only the animation flag changes. No layout shift, so the compositing win is real without moving anything.

In the worktree: `session-dot-state` 24/24 pass (including the 4 new `$runningArcsAnimated` checks) and `compositing.test.mjs` 6/6 pass on the trace aggregation. Implementation matches the title: compositing is now measured and the sidebar no longer grows its animated layers.

Heads up: this PR is stacked on `5cc1369fa` which has diverged from current `main`, so it will need a rebase before it can merge cleanly. The fix itself is ready.

Issue #76577, opened 2026-08-02
[Bug] docker backend cannot read attached binary files — host path not mounted into sandbox


STRML:
Environment: desktop app, `terminal.backend: docker`.

## Problem
Pasting/dropping a **binary** file (e.g. `.zip`) hands the model a host path (`/Users/<user>/three-rings.zip`) instead of inlined text. Under the docker backend that path dangles — the container has its own filesystem and the host path is not mounted, so the agent cannot `unzip`/read the attachment at all. Text files work only because they are inlined into the message.

## Root cause (traced)
- Desktop OS drop of a non-image file → `@file:` ref. On submit, `uploadComposerAttachment` (`apps/desktop/src/app/session/hooks/use-prompt-actions/index.ts:106`) sets `uploadBytes = remote || attachmentPathNeedsUpload(path, backendCwd)`.
- `attachmentPathNeedsUpload` (same file, ~line 87) only returns true for a **Windows absolute path + POSIX backend**. macOS host → docker backend returns **false**, so the raw host path is sent and bytes never uploaded.
- `agent/context_references.py::_binary_reference_block` (~line 575) renders that host path verbatim into the prompt. The container cannot see it.

## Existing machinery (reuse it)
`tools/credential_files.py` already provides host↔container path translation and bind-mounts cache dirs into docker sandboxes:
- `_CACHE_DIRS` (mount list incl. `cache/images`, `cache/documents`, flat `images/` from #69575)
- `to_agent_visible_cache_path` / `map_cache_path_to_container`
- `from_agent_visible_cache_path`
Image generation and #69575 already route through this.

## Proposed fix (generalizes the #69575 mount)
1. **Broaden `attachmentPathNeedsUpload`** to also return true when the backend is a container (docker/ssh/etc.), routing every OS drop through the existing `file.attach` data_url pipeline.
2. **Stage into a bind-mounted dir**: add an `attachments` entry to `_CACHE_DIRS` (same pattern as the `images/` mount in #69575) and point the desktop attachment staging dir (`_desktop_attachment_dir` in `tui_gateway/server.py`) there, so staged files land inside the container.
3. **Translate at render**: in `_expand_file_reference`/`_binary_reference_block`, apply the existing `to_agent_visible_cache_path(...)` so the model gets the in-container path instead of the host path.

Tiny, reuses all existing plumbing, and removes the text/binary asymmetry.

## Related
- #69575 (image-only mount, closed/fixed)
[...]

PR #98646, opened 2026-08-30
fix(gateway): allowlist session attachment staging dir in @file: ref expansion

### LINKED ISSUE #98634: [Bug]: file.attach returns @file: refs that prompt expansion always refuses (allowed_root=cwd vs attachments/ staging dir)
### What happens

`file.attach` stages an uploaded file into the profile home's `attachments/` dir and returns a `ref_text` like `@file:/home/u/.hermes/attachments/notes.txt` for the client to include in the prompt. But when that prompt is submitted, `preprocess_context_references` (called from `_run_prompt_submit` with `allowed_root=cwd`) resolves the ref against the session workspace and **refuses it**: `path is outside the allowed workspace`. The file's content is never inlined (text files) and no binary pointer block is emitted (binary files) — the model just sees the raw `@file:` token plus a warning block.

So the gateway's own attach RPC hands out a reference that the gateway's own expansion then rejects, for every staged file that isn't already inside the workspace (i.e. the entire remote-upload case the RPC exists for, per its docstring).

The agent can still reach the file by tool-reading the absolute path, so attachments aren't fully broken — but the intended inline expansion (`📄` text blocks / `📎` binary pointer blocks) never happens for staged attachments, and every such prompt is persisted with a `--- Context Warnings ---` block.

### Reproduction (hermes-agent v0.20.6, `hermes serve`, token mode)

Over one WS connection:

1. `session.create`
2. `file.attach` with `data_url: "data:text/plain;base64,…"`, `name: "probe.txt"` → returns `"ref_text": "@file:/Users/<user>/.hermes/attachments/probe.txt"` (absolute — the staging dir is outside the session cwd, so `_attachment_ref_path` falls back to the absolute path)
3. `prompt.submit` with `text: "caption line\n" + ref_text`

Persisted user row (verbatim, via REST after the turn):

```
caption line
@file:/Users/<user>/.hermes/attachments/probe.txt

--- Context Warnings ---
- @file:/Users/<user>/.hermes/attachments/probe.txt: path is outside the allowed workspace
```

Same result for a binary upload (no `📎` block, same warning). Control: a hand-typed `@file:` ref pointing **inside** the workspace expands normally (`--- Attached Context ---` with the `📄` inline block / `📎` binary pointer), confirming the refusal is specifically the allowed-root check, not expansion generally.

### Where the pieces disagree

- `_stage_session_file_attachment` (tui_gateway/server.py) deliberately stages into the profile home's `attachments/` dir — the docstring notes it must land there so container bind mounts can see it (#76577).
- `_attachment_ref_path` returns a workspace-relative path when possible, "or the absolute path if outside" — staged uploads are always outside, so the ref is always absolute.
- `_run_prompt_submit` calls `preprocess_context_references(..., allowed_root=cwd)`, and `_resolve_path` raises `path is outside the allowed workspace` for any resolved path not under `allowed_root`.

Each piece is individually reasonable (the allowed-root check exists because untrusted remote text reaches expansion), but composed they mean `file.attach` + `ref_text` + `prompt.submit` can never inline a staged attachment.

### Suggested direction

Allowlist the session's own staged-attachment directory (the `_desktop_attachment_dir` for the session's profile) in the expansion's allowed roots — it holds only files the client explicitly uploaded through the authenticated `file.attach` RPC, so inlining them doesn't widen what untrusted message text can exfiltrate (`_ensure_reference_path_allowed`'s credential deny-list would still apply). Alternatively, have `file.attach` return refs the check accepts (e.g. stage a workspace-visible link) — but the allowlist seems smaller.

### Context

Found while implementing file attachments in a third-party desktop-contract client (contract 6). Client-side behavior is fine either way (the model tool-reads the file), so this is about the lost inline expansion + warning noise in persisted history, not a hard failure.

liuhao1024:
## What does this PR do?

`file.attach` stages uploaded files into the profile home's `attachments/` dir — deliberately **outside** the session workspace so container bind mounts can see them (#76577) — and hands back absolute `@file:` refs for the client to include in the prompt. But `_run_prompt_submit` expands refs with a single `allowed_root` (the session cwd), so `_resolve_path` refused the gateway's own refs (`path is outside the allowed workspace`): staged attachments never inlined (`📄`/`📎` blocks) and every such prompt was persisted with a `--- Context Warnings ---` block.

This follows the allowlist direction from the issue: `preprocess_context_references(_async)` now accepts a **sequence** of allowed roots (a resolved path under ANY root passes; a single root value keeps working; `None`/empty still defaults to the cwd boundary), and `_run_prompt_submit` passes `(cwd, _desktop_attachment_dir(session))`. The staging dir holds only files uploaded through the authenticated `file.attach` RPC, so this does not widen what untrusted message text can reach, and the credential deny-list in `_ensure_reference_path_allowed` still runs on the resolved path. Staging-dir resolution is wrapped so a failure there can never break the submit itself.

## Related Issue

Fixes #98634

## Type of Change

## Changes Made

- `agent/context_references.py` — `allowed_root` now accepts a single root or a sequence; new `_normalize_allowed_roots` helper; `_resolve_path` accepts a path under any of the allowed roots (same refusal message otherwise). Internal signatures thread `tuple[Path, ...]`.
- `tui_gateway/server.py` — `_run_prompt_submit` passes `[cwd, _desktop_attachment_dir(session)]` as the allowed roots, with a guard so staging-dir resolution failures degrade to the previous single-root behavior.
- `tests/agent/test_context_references.py` — regression tests: a secondary root's file inlines while a path under neither root is still refused; the credential deny-list still fires for sensitive files inside an allowed secondary root.
- `tests/test_tui_gateway_server.py` — end-to-end wiring test: a `prompt.submit` with a staged-attachment `@file:` ref inlines the file body, while an unrelated outside path is still refused.

## How to Test

1. `python -m pytest tests/agent/test_context_references.py -q` — 15 passed (includes the two new multi-root regressions).
2. `python -m pytest tests/test_tui_gateway_server.py -q` — 620 passed (includes the new staged-attachment submit test).
3. `python -m pytest tests/agent/test_context_refs_concurrent.py tests/agent/test_plugin_context_references.py tests/cli/test_cli_codex_context_reference.py tests/gateway/test_context_ref_expansion_runtime.py tests/tools/test_send_message_tool.py -q` — 83 passed.
4. Manual reproduction per the issue (observed on the fix): `session.create` → `file.attach` with a `data_url` → `prompt.submit` with the returned `ref_text`; the persisted user row now contains `--- Attached Context ---` with the inline block instead of the `path is outside the allowed workspace` warning.

### Code

### DIFF
--- a/agent/context_references.py
+++ b/agent/context_references.py
@@ -9,7 +9,7 @@ import re
 import subprocess
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Awaitable, Callable
+from typing import Awaitable, Callable, Sequence
 
 from agent.model_metadata import estimate_tokens_rough
 from hermes_cli._subprocess_compat import IS_WINDOWS, windows_hide_flags
@@ -215,7 +215,7 @@ def preprocess_context_references(
     cwd: str | Path,
     context_length: int,
     url_fetcher: Callable[[str], str | Awaitable[str]] | None = None,
-    allowed_root: str | Path | None = None,
+    allowed_root: str | Path | Sequence[str | Path] | None = None,
 ) -> ContextReferenceResult:
     coro = preprocess_context_references_async(
         message,
@@ -242,7 +242,7 @@ async def preprocess_context_references_async(
     cwd: str | Path,
     context_length: int,
     url_fetcher: Callable[[str], str | Awaitable[str]] | None = None,
-    allowed_root: str | Path | None = None,
+    allowed_root: str | Path | Sequence[str | Path] | None = None,
 ) -> ContextReferenceResult:
     refs = parse_context_references(message)
     if not refs:
@@ -250,10 +250,10 @@ async def preprocess_context_references_async(
 
     cwd_path = Path(cwd).expanduser().resolve()
     # Default to the current working directory so @ references cannot escape
-    # the active workspace unless a caller explicitly widens the root.
-    allowed_root_path = (
-        Path(allowed_root).expanduser().resolve() if allowed_root is not None else cwd_path
-    )
+    # the active workspace unless a caller explicitly widens the root. A
+    # sequence widens the boundary to several roots at once (e.g. the
+    # gateway's workspace plus the session's attachment staging dir, #98634).
+    allowed_roots = _normalize_allowed_roots(allowed_root) or (cwd_path,)
     warnings: list[str] = []
     blocks: list[str] = []
     injected_tokens = 0
@@ -270,7 +270,7 @@ async def preprocess_context_references_async(
                 ref,
                 cwd_path,
                 url_fetcher=url_fetcher,
-                allowed_root=allowed_root_path,
+                allowed_root=allowed_roots,
             )
             for ref in refs
         )
@@ -330,7 +330,7 @@ async def _expand_reference(
     cwd: Path,
     *,
     url_fetcher: Callable[[str], str | Awaitable[str]] | None = None,
-    allowed_root: Path | None = None,
+    allowed_root: tuple[Path, ...] | None = None,
 ) -> tuple[str | None, str | None]:
     try:
         if ref.kind == "file":
@@ -369,7 +369,7 @@ def _expand_file_reference(
     ref: ContextReference,
     cwd: Path,
     *,
-    allowed_root: Path | None = None,
+    allowed_root: tuple[Path, ...] | None = None,
 ) -> tuple[str | None, str | None]:
     path = _resolve_path(cwd, ref.target, allowed_root=allowed_root)
     _ensure_reference_path_allowed(path)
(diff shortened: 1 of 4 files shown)


Enough1122:

Allows `@file:` refs to resolve inside both workspace and attachment staging dir without widening to arbitrary filesystem.

- **API (`agent/context_references.py:9+`):** `allowed_root` now accepts `str|Path|Sequence[str|Path]`, normalized via `_normalize_allowed_roots` to `tuple[Path,...]` with `resolve()`. `preprocess_context_references_async` and `_expand_{file,folder}_reference` propagate tuple.
- **Security check (`:489+`):** `_resolve_path` now checks `any(resolved.is_relative_to(root) for root in allowed_root)`. Previously `relative_to` single root; new `any` semantics correctly widens to union of roots but never narrows. Staging dir still bounded, not `cwd_path.parent`.
- **Test coverage (`tests/agent/test_context_references.py`):** Expect assert for sequence handling and gateway staging dir inclusion (issue #98634 context). Good.

Non-blocking:
- `_normalize_allowed_roots` returns `None` when input is `None`, caller does `or (cwd_path,)` — fallback remains single cwd. If caller passes empty sequence `()`, `tuple()` is falsy and falls back to cwd too, which is safe (empty widening is no-op). Consider explicitly documenting empty sequence behavior.

liuhao1024:
Thanks for the review — fair point on the empty-sequence behavior: the falsy-tuple fallback to `cwd_path` makes an empty widening a no-op, which fails safe. Noted for a docstring touch-up if this API grows more callers.

PR #98653, opened 2026-08-30
fix(context): allow staged file.attach refs to expand inline

### LINKED ISSUE #98634: [Bug]: file.attach returns @file: refs that prompt expansion always refuses (allowed_root=cwd vs attachments/ staging dir)

kokhlo:
## What

`file.attach` stages uploads into the session profile's `attachments/` dir and hands back an absolute `@file:` ref — but `preprocess_context_references` was called with `allowed_root=cwd` only, so every staged attachment (i.e. every remote upload, the exact case the RPC exists for) died with `path is outside the allowed workspace`: no `📄` inline block, no `📎` binary pointer, just a `--- Context Warnings ---` block persisted into history.

## Root cause

Each piece was individually fine, composed they contradicted each other:

- `_stage_session_file_attachment` deliberately stages outside the workspace (bind-mount visibility, #76577), and `_attachment_ref_path` correctly returns an absolute ref for it;
- `_resolve_path` accepted exactly one allowed root, and all callers passed the session cwd.

## Fix

Opt-in widening, as suggested in the issue:

- `preprocess_context_references(_async)` and the file/folder expanders take `extra_allowed_roots`; `_resolve_path` now accepts a path under the primary root **or** any extra root;
- `_run_prompt_submit` passes the session's `_desktop_attachment_root` (its own staged-attachment dir — pure path computation, no `mkdir` on the submit path) as the extra root;
- the credential deny-list is unchanged and still runs on the resolved path afterwards: a secret-bearing file staged inside `attachments/` (e.g. `.env`) is still refused.

Only files that arrived through the authenticated `file.attach` RPC can live in that dir, so this doesn't widen what untrusted remote text can reach — the workspace boundary for arbitrary paths is intact (verified: an unrelated outside-workspace path is still refused).

## Testing

- 3 new unit tests in `tests/agent/test_context_references.py`: staged file expands with the extra root; the same ref is still refused without it; the credential deny-list still fires inside the extra root.
- New gateway test: `prompt.submit` now passes `extra_allowed_roots` containing the session's `attachments/` dir.
- Full local suites green: `test_context_references.py`, `test_context_refs_concurrent.py`, `test_plugin_context_references.py`, `test_context_ref_expansion_runtime.py` (34 passed), `test_tui_gateway_server.py -k "attach or context or prompt_submit"` (95 passed), including both `prompt.submit` context-expansion tests.
- End-to-end repro of the issue scenario (workspace cwd + staging dir under the profile home): before → `path is outside the allowed workspace`, no inline content; after → `📄` block with the file content, zero warnings; `attachments/.env` still refused by the deny-list; unrelated outside path still refused.

Fixes #98634

### DIFF
--- a/agent/context_references.py
+++ b/agent/context_references.py
@@ -216,6 +216,7 @@ def preprocess_context_references(
     context_length: int,
     url_fetcher: Callable[[str], str | Awaitable[str]] | None = None,
     allowed_root: str | Path | None = None,
+    extra_allowed_roots: tuple[str | Path, ...] | list[str | Path] | None = None,
 ) -> ContextReferenceResult:
     coro = preprocess_context_references_async(
         message,
@@ -223,6 +224,7 @@ def preprocess_context_references(
         context_length=context_length,
         url_fetcher=url_fetcher,
         allowed_root=allowed_root,
+        extra_allowed_roots=extra_allowed_roots,
     )
     # Safe for both CLI (no loop) and gateway (loop already running).
     try:
@@ -243,6 +245,7 @@ async def preprocess_context_references_async(
     context_length: int,
     url_fetcher: Callable[[str], str | Awaitable[str]] | None = None,
     allowed_root: str | Path | None = None,
+    extra_allowed_roots: tuple[str | Path, ...] | list[str | Path] | None = None,
 ) -> ContextReferenceResult:
     refs = parse_context_references(message)
     if not refs:
@@ -253,6 +256,9 @@ async def preprocess_context_references_async(
     # the active workspace unless a caller explicitly widens the root.
     allowed_root_path = (
         Path(allowed_root).expanduser().resolve() if allowed_root is not None else cwd_path
+    )
+    extra_allowed_root_paths = tuple(
+        Path(root).expanduser().resolve() for root in (extra_allowed_roots or ())
     )
     warnings: list[str] = []
     blocks: list[str] = []
@@ -271,6 +277,7 @@ async def preprocess_context_references_async(
                 cwd_path,
                 url_fetcher=url_fetcher,
                 allowed_root=allowed_root_path,
+                extra_allowed_roots=extra_allowed_root_paths,
             )
             for ref in refs
         )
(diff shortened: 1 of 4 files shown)


alt-glitch:

Duplicate of #98646 — both repair the same staged file.attach @file expansion boundary while retaining the credential deny-list. #98646 is the earlier, broader implementation.

PR #98656, opened 2026-08-30
fix(desktop): attach err.statusCode in fetchJson so mint 401s classify as reauth

### LINKED ISSUE #98647: Desktop: a 401 from the WS-ticket mint is reported to the user as "Could not reach the remote Hermes gateway"
**Component:** `apps/desktop` (Electron main)
**Severity:** low code impact, high diagnostic cost
**Found on:** desktop built from `main` 2026-08-30; gateway `v0.20.6` (`release_date` 2026.8.27), `auth_providers: ["basic"]`, `auth_flows: ["cookie","native_pkce"]`

## Summary

When the gateway rejects a WebSocket-ticket mint with **401**, the desktop reports it as a
*connectivity* failure rather than an *auth* failure. The user is told the gateway is
unreachable, and the one instruction that would actually fix it — "Open Settings → Gateway
and sign in again" — is the branch that never runs.

The status code is formatted into the error **message string** but never attached as a
property, and the classifier reads the property.

## Reproduce

1. Connect the desktop to a gated gateway in `oauth` mode and sign in (native PKCE).
2. Restart the gateway with no `dashboard.basic_auth.secret` configured, so the basic-auth
   plugin regenerates its per-process signing key and all issued sessions become invalid.
   (Any cause of a 401 on the mint reproduces it; this is just the easy one.)
3. Relaunch the desktop.

**Observed** — `~/.hermes/logs/desktop.log`:

```
[boot] Desktop boot failed: Could not reach the remote Hermes gateway while refreshing its WebSocket ticket. Try reconnecting.
[renderer console:main] [profiles] refreshProfiles failed after 3 attempt(s): Error: Error invoking remote method 'hermes:api': Error: Could not reach the remote Hermes gateway while refreshing its WebSocket ticket. Try reconnecting.
```

The gateway was healthy and reachable throughout — `GET /login` 200, `GET /api/health` 200,
`GET /api/status` 200 — and `POST /api/auth/ws-ticket` was returning a well-formed 401:

```json
{"error":"unauthenticated","detail":"Unauthorized","reason":"no_cookie","login_url":"/login"}
```

**Expected:** the auth branch — "Reached the gateway over HTTP, but the OAuth session was
rejected… Open Settings → Gateway and sign in again." — plus `needsOauthLogin: true` so the
renderer can surface the re-auth prompt.

## The app already gets this right when there is NO session — only the rejected-session path is broken

Worth stressing, because it narrows the bug. With no stored session at all, the pre-mint
guard fires and the message is exactly right:

```
[boot] Desktop boot failed: Remote Hermes gateway uses OAuth, but you are not signed in.
Open Settings → Gateway and click "Sign in", or switch back to Local.
```

That is `makeUnsignedOauthError()`, reached via the `oauthSessionIsLive(...) &&
oauthGuardMayHardFail(...)` check that runs *before* `mintGatewayWsTicket`. So the desktop is
perfectly capable of telling the user to sign in.

The failure is specifically **"a session exists but the gateway rejects it"** — the guard
sees a stored native session, passes, and the 401 then arrives from the mint, where the
classification is broken. That is also the far more common real-world case: sessions expire,
get revoked, or (as here) are invalidated wholesale by a gateway restart. The path that
handles the *routine* failure is the one that misreports it.

## Root cause

`apps/desktop/electron/main.ts`, in `fetchJson` (~line 5053):

```js
if ((res.statusCode || 500) >= 400) {
  reject(new Error(`${res.statusCode}: ${text || res.statusMessage}`))
  return
}
```

The status lands in the message text only. Nothing sets `err.statusCode`.

`apps/desktop/electron/connection-config.ts` (~line 119) classifies by property:

```js
function isGatewayAuthRejection(error) {
  if (error && typeof error === 'object' && error.needsOauthLogin === true) return true
  const statusCode = Number(error && typeof error === 'object' ? error.statusCode : NaN)
  return statusCode === 401 || statusCode === 403
}
```

`Number(undefined)` is `NaN`, so a 401 classifies as *not* an auth rejection, and
`gatewayTicketFailure` (~line 129) selects the transport message.

Two things in the surrounding code show the property is meant to exist:

- `connection-config.ts:137-147` explicitly preserves `error.statusCode` through the wrapper,
  with a comment that downstream Cloud classification depends on it surviving.
- `makeNousCloudBackendDownError` / `isServerSideHttpError` also key off `statusCode`, so
  **502/503/504 Cloud-down detection is silently dead on this path too** — same missing
  property, and that one is not just a wording problem.

Secondary effect: `withTransientRetries` retries anything that isn't classified as an auth
rejection, so every dead-session mint is retried 3× before failing. Harmless, but it is
hammering a session that cannot recover.

## Suggested fix

Attach the status in the fetch layer, where it is already known:

```js
if ((res.statusCode || 500) >= 400) {
  const err = new Error(`${res.statusCode}: ${text || res.statusMessage}`)
  err.statusCode = res.statusCode
  reject(err)
  return
}
```

Everything downstream already reads it; no other call site needs to change.

Please check the sibling path too — `fetchJsonViaOauthSession` (~line 7350) feeds the same
`mintGatewayWsTicket` when there is no native bearer, and needs the same property for the
cookie-mode case to classify correctly.

A regression test would be worthwhile: `backend-health.test.ts` already asserts
`isReauthRequiredError(new Error('Could not reach the remote Hermes gateway')) === false`,
which locks in the *message-based* view; a test that a 401 from `fetchJson` produces
`needsOauthLogin` would have caught this.

liuhao1024:
## What does this PR do?

When a gated gateway rejects a WebSocket-ticket mint with 401 (expired, revoked, or wholesale-invalidated session), the desktop reported it as a *connectivity* failure — "Could not reach the remote Hermes gateway" — instead of the actionable "sign in again" branch, and `withTransientRetries` retried the dead session 3× before giving up.

Root cause: `fetchJson` (`apps/desktop/electron/main.ts`) formats the HTTP status into the reject **message** but never sets `err.statusCode`. Every downstream classifier reads the structured property: `isGatewayAuthRejection` (connection-config.ts) does `Number(error.statusCode)` → `NaN` → "not an auth rejection", and `isServerSideHttpError` / `makeNousCloudBackendDownError` have the same blind spot on this path for 502/503/504. The sibling `fetchJsonViaOauthSession` already sets the property (and its cookie-mode 401s classify correctly); the comments in both `backend-health.ts` and `connection-config.ts` document "the fetch layer attaches err.statusCode" as the contract — `fetchJson` just never fulfilled it.

The fix attaches the property at the fetch layer, where the status is already known. No call site changes: everything downstream already reads it. Verified `fetchPublicJson` (the other function with a bare `reject(new Error(...))`) was deliberately left alone — its consumers either degrade via `catch` using only `error.message` or classify via the legacy `/^40[13]:/` message prefix, so nothing downstream of it reads the property.

## Related Issue

Fixes #98647

## Type of Change

## Changes Made

- `apps/desktop/electron/main.ts` — `fetchJson`'s 4xx/5xx branch now attaches `err.statusCode = res.statusCode || 500` to the rejected error (mirrors `fetchJsonViaOauthSession`), so 401/403 route to reauth and 502/503/504 reach the Cloud-down classifier.
- `apps/desktop/electron/connection-config.test.ts` — regression test pinning the exact error shape `fetchJson` produces (message prefix + structured `statusCode`) through `gatewayTicketFailure`, asserting the auth branch and `needsOauthLogin: true`.

## How to Test

1. `cd apps/desktop && npx vitest run --project electron electron/connection-config.test.ts`
   - Observed result: 116 passed, including the new `ticket-mint 401 from fetchJson routes to reauth via structured statusCode (#98647)`.
2. `cd apps/desktop && npx vitest run --project electron`
   - Observed result: 1959 passed; the single failure (`electron/managed-ssh-update.test.ts`, exit 127) reproduces on a clean `upstream/main` checkout with this change stashed — pre-existing, unrelated.
3. `npx tsc --noEmit -p tsconfig.electron.json` — passes.
4. Behavioral check (from the issue): with a stored native session invalidated by a gateway restart, the WS-ticket mint now rejects with a 401 carrying `statusCode: 401` → `gatewayTicketFailure` selects the auth message ("Reached the gateway over HTTP, but the OAuth session was rejected… Open Settings → Gateway and sign in again") and sets `needsOauthLogin: true`, so the renderer surfaces the re-auth prompt instead of the unreachable-gateway message, and `withTransientRetries` fails fast instead of hammering the dead session.

### Code

- [x] I've run `pytest tests/ -q` and all tests pass (change is desktop-TS only; the equivalent suite `vitest run --project electron` passes except one pre-existing failure unrelated to this change, reproduced on clean main)
- [x] I've tested on my platform: macOS 15 (arm64, Darwin 25.4.0)

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A (code comment added at the fixed branch documenting the classifier contract)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — N/A (Node fetch layer, platform-independent)

## Screenshots / Logs

Not applicable — behavior verified via unit tests (see How to Test).

### DIFF
--- a/apps/desktop/electron/connection-config.test.ts
+++ b/apps/desktop/electron/connection-config.test.ts
@@ -1342,3 +1342,25 @@ test('OAuth ticket-mint 401 stays on the reauth path (never Cloud-down)', () => 
   assert.equal((wrapped as any).needsOauthLogin, true)
   assert.equal((wrapped as any).statusCode, 401)
 })
+
+// Regression (#98647): fetchJson (electron/main.ts) attaches err.statusCode on
+// 4xx/5xx responses. Before that, a dead-session 401 from the WS-ticket mint
+// carried the code only in the message text; isGatewayAuthRejection read
+// Number(undefined) === NaN and every expired/revoked session was reported as
+// "Could not reach the remote Hermes gateway" (and retried by
+// withTransientRetries, hammering a session that cannot recover). This pins
+// the exact error shape fetchJson produces — message prefix AND structured
+// statusCode — so the fetch-layer → classifier contract cannot silently
+// regress again.
+test('ticket-mint 401 from fetchJson routes to reauth via structured statusCode (#98647)', () => {
+  const ticketErr = new Error(
+    '401: {"error":"unauthenticated","detail":"Unauthorized","reason":"no_cookie","login_url":"/login"}'
+  ) as any
+  ticketErr.statusCode = 401
+
+  const wrapped = gatewayTicketFailure(ticketErr, 'auth message', 'transport message') as any
+
+  assert.equal(wrapped.message, 'auth message')
+  assert.equal(wrapped.needsOauthLogin, true)
+  assert.equal(wrapped.statusCode, 401)
+})
--- a/apps/desktop/electron/main.ts
+++ b/apps/desktop/electron/main.ts
@@ -5052,7 +5052,13 @@ function fetchJson(url, token, options: any = {}) {
               const text = Buffer.concat(chunks).toString('utf8')
 
               if ((res.statusCode || 500) >= 400) {
-                reject(new Error(`${res.statusCode}: ${text || res.statusMessage}`))
+                // Downstream classification (isGatewayAuthRejection /
+                // isServerSideHttpError) reads err.statusCode, never the
+                // "401: ..." message prefix — same contract as
+                // fetchJsonViaOauthSession below.
+                const err = new Error(`${res.statusCode}: ${text || res.statusMessage}`) as any
+                err.statusCode = res.statusCode || 500
+                reject(err)
 
                 return
               }


alt-glitch:

Related to #96449, which makes the same fetch-layer status-code repair more broadly by also covering fetchPublicJson. This PR is the focused fetchJson subset for #98647.

liuhao1024:
Thanks @alt-glitch for the cross-link — confirmed. #96449 fixes the identical root cause (`fetchJson` rejecting with the status only in the message string, so `isGatewayAuthRejection()` reads `Number(undefined)` and a dead-session 401 lands on the transport-failure path): the `main.ts` hunk here is the same line-level repair, and #96449 additionally covers `fetchPublicJson` and was opened three days earlier (Aug 27 vs today).

For timeline context: this PR was submitted against the freshly-filed #98647 before the triage cross-link existed, so the overlap wasn't visible at claim time. Now that it is, I'm withdrawing in favor of #96449 — it is the superset fix and already carries equivalent behavioral coverage (401 → `needsOauthLogin`, with the pre-fix message-only shape documented as the guarded regression), so there is no unique delta here worth a second open PR.

Closing this one. Hope #96449 lands soon — the "gateway unreachable" misreport cost users real debugging time.

PR #30324, opened 2026-05-22
fix(achievements): detect Ollama native name:tag format in is_local_model_name()

sprmn24:
## What does this PR do?

`is_local_model_name()` checks a fixed list of string markers (`ollama`, `llama.cpp`, `localhost`, etc.) but Ollama's native model strings are bare `name:tag` format — e.g. `qwen2.5:3b`, `llama3:8b`, `mistral:7b`. None of these contain any of the existing markers, so the function returned `False` and achievements like **Open Weights Pilgrim** never fired even when the user genuinely ran a local model.

## Root cause

Traced via `state.db`: Ollama sessions record `model='qwen2.5:3b'` with `billing_provider=''` — confirming genuine local inference. The string `qwen2.5:3b` doesn't match any existing marker, so `is_local_model_name()` returned `False`.

This is exactly the second case mentioned in the #27652 review: *'`is_local_model_name()` doesn't match the model string Hermes recorded'*.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Changes Made

Added a second return path in `is_local_model_name()`: if the model string contains no `/` (no provider prefix) and matches `^[a-z0-9._-]+:\d+\.?\d*[bBmMkK]$`, treat it as an Ollama native `name:tag` format and return `True`.

## How to Test

```bash
sqlite3 ~/.hermes/state.db "SELECT id, model, billing_provider FROM sessions WHERE model LIKE '%qwen2.5%'"
# Shows: qwen2.5:3b with billing_provider=''

python3 -m pytest tests/plugins/test_achievements_plugin.py -v
# 15 passed
```

## Checklist

- [x] My code follows the project style guidelines
- [x] I have added tests that prove my fix is effective
- [x] All existing tests pass locally

PR #98659, opened 2026-08-30
fix(achievements): detect local sessions by route

ataraxiaone:
## What does this PR do?

Make **Open Weights Pilgrim** count local-model sessions from the session's recorded route metadata, not only substrings in the model name.

Hermes can record a local LM Studio or OpenAI-compatible session with a plain model identifier such as `qwen3.6-27b-mtp`. That name contains none of the achievement plugin's existing local markers, even though the session row records `billing_provider=lmstudio` and/or a loopback `billing_base_url`. The scan therefore treated genuine local inference as remote inference.

## Changes

- Treat known local providers (`lmstudio`, `ollama`, `llama.cpp`, `local`, and local vLLM aliases) as local-model evidence.
- Treat `localhost` and IP loopback base URLs as local-model evidence.
- Include provider and base-URL route fields in the session fingerprint so checkpoint reuse cannot preserve stale route classification.
- Rehydrate route metadata from the current session row even when message-derived statistics come from the checkpoint cache.
- Add unit and scan-path regression coverage, including a negative control proving that a remote Qwen session does not count merely because the model is open-weight.

## Relationship to #30324

This complements rather than duplicates #30324. That PR recognizes Ollama's native `name:tag` model strings. This PR handles cases where the model name itself is ambiguous and the authoritative evidence is the recorded provider or loopback route, including LM Studio and generic local OpenAI-compatible servers.

## Verification

Rebased onto current `main`, then ran:

```bash
scripts/run_tests.sh \
  tests/plugins/test_achievements_plugin.py \
  plugins/hermes-achievements/tests/test_achievement_engine.py
python3 -m py_compile plugins/hermes-achievements/dashboard/plugin_api.py
```

Result: **19 passed, 0 failed**; Python compilation passed.

The fix was also exercised against an actual Hermes session database containing LM Studio-routed sessions: the scanner recognized the local sessions and unlocked Open Weights Pilgrim without counting remote Qwen sessions.

## Risk

Low and confined to the dashboard-only achievements plugin. No model-visible tools, provider routing, or inference behavior changes. Loopback detection uses parsed hostnames and `ipaddress.is_loopback`; non-loopback remote endpoints remain excluded.

### DIFF
--- a/plugins/hermes-achievements/dashboard/plugin_api.py
+++ b/plugins/hermes-achievements/dashboard/plugin_api.py
@@ -4,6 +4,7 @@ Mounted at /api/plugins/hermes-achievements/ by Hermes dashboard.
 """
 from __future__ import annotations
 
+import ipaddress
 import json
 import math
 import re
@@ -11,6 +12,7 @@ import threading
 import time
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set
+from urllib.parse import urlparse
 
 try:
     from hermes_constants import get_hermes_home
@@ -258,6 +260,8 @@ def session_fingerprint(meta: Dict[str, Any]) -> Dict[str, Any]:
         "last_active": meta.get("last_active"),
         "started_at": meta.get("started_at"),
         "model": meta.get("model"),
+        "billing_provider": meta.get("billing_provider"),
+        "billing_base_url": meta.get("billing_base_url"),
         "title": meta.get("title") or meta.get("preview") or "Untitled",
     }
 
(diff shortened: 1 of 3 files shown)


Enough1122:

Fixes local-model detection for achievements by considering provider and base_url routing, not just model name substring.

- **New helpers (`plugins/hermes-achievements/dashboard/plugin_api.py:342+`):** `is_local_model_provider` normalizes `provider_name` (lower, `_`→`-`) against `lmstudio/lm-studio/ollama/llamacpp/...`; `is_local_model_base_url` parses hostname via `urlparse`, checks `localhost` or `ip.is_loopback`; `is_local_model_session` ORs all three. Covers routed local models where `model_name` alone is ambiguous.
- **Persistence fix (`:694+`):** `scan_sessions` now overwrites `model_names/model_providers/model_base_urls` from `meta` on each scan, even when message analysis came from checkpoint. Prevents stale cached routing surviving provider correction. Previous `setdefault` logic removed.
- **Fingerprint (`:258+`):** `session_fingerprint` includes `billing_provider/base_url` so checkpoint invalidation triggers on routing change.
- **Tests:** `test_achievement_engine.py` added case for provider/base_url local detection.

Non-blocking:
- `is_local_model_base_url` returns False on `ValueError` from urlparse — good. If hostname is `127.0.0.1` with non-default port, `ip.is_loopback` still true, correctly flagged as local.

PR #98660, opened 2026-08-30
fix(desktop): store transparent pre-paint background when glass/vibrancy is active (macOS dark mode white flash)

### LINKED ISSUE #98561: [Bug]: macOS Desktop — opening the app is still blinding white in dark mode and dark theme
## Bug Description

Opening Hermes Desktop on macOS still paints a **blinding white** first frame (and it can linger) even when:

- macOS itself is in Dark Mode, and
- the app appearance is an explicit Dark theme / dark skin.

This has been reported for over a month and is still present. The existing anti-flash work (`show: false` until `ready-to-show`, `index.html` pre-paint, persisted `nativeTheme.themeSource`, `getWindowBackgroundColor()`) is being defeated by the macOS vibrancy / glass first-frame path. Night use is painful.

This is **not** a stuck blank/white renderer after a failed remote OAuth boot. The app does come up. The failure is the native window + first compositor frame being light/white on every open.

## Steps to Reproduce

1. On macOS, set System Settings → Appearance → Dark.
2. In Hermes Desktop, set appearance to Dark (not only “system”) and pick a dark theme/skin.
3. Fully quit Hermes Desktop (Cmd+Q), not just hide.
4. Reopen Hermes Desktop from the Dock / app icon.
5. Repeat: new window (New Window) and cold launch after reboot.

## Expected Behavior

From the first visible pixel, the window is dark. No white or near-white flash, even for a frame. Dark OS + dark app theme must never paint a light NSVisualEffectView or a UA-white webContents backing.

## Actual Behavior

The window opens **blinding white** (full window, not a small chrome glitch) despite Dark Mode and a dark theme. The themed UI only appears after. Happens on every open.

## Affected Component

Hermes Desktop (`apps/desktop`) on macOS. Template dropdown has no Desktop option.

- `apps/desktop/electron/main.ts` — `chatWindowSurfaceOptions`, `createWindow`, `wireWindowReveal`, `getWindowBackgroundColor`, `applyWindowTranslucency`
- `apps/desktop/electron/translucency.ts` — `windowBackingOptions`
- `apps/shared/src/translucency.ts` — `defaultTranslucencyState`, `glassActive`, `vibrancyFor`, `DEFAULT_VALUES`
- `apps/desktop/index.html` — pre-paint script
- `apps/desktop/src/themes/context.tsx` — `applyTheme` / `hermes-boot-background`

## Debug Report

N/A — pure Desktop UI first-paint / native window. No crash, no gateway traceback. Do not require `hermes debug share`.

## Operating System

macOS

## Python Version

N/A (Desktop UI)

## Hermes Version

Hermes Desktop, macOS. Still present on current `main` (grounded against `66666f6e2eca`).

UI-only. No debug paste.

## Root Cause Analysis

macOS chat windows **always** get a vibrancy material, and glass **omits** the opaque themed `backgroundColor` that the anti-flash comments say is required.

| What | File | Symbol / detail |
|------|------|-----------------|
| Vibrancy on every chat window | `apps/desktop/electron/main.ts` | `chatWindowSurfaceOptions()` sets `vibrancy: IS_MAC ? vibrancyForTranslucency(translucencyState)` |
| Vibrancy ignores `backgroundColor` (already documented in-tree) | `apps/desktop/electron/main.ts` ~L921–925, ~L13950–13952 | “macOS paints an NSVisualEffectView that … ignores `backgroundColor` — so a dark-themed app … flashes a white material” |
| Glass omits backing | `apps/desktop/electron/translucency.ts` `windowBackingOptions` | `return glassActive(state) ? {} : { backgroundColor: themedColor }` |
| Glass ships on | `apps/shared/src/translucency.ts` | “Glass ships ON”; mac dark default `{ intensity: 22, … material: 'titlebar' }` |
| `glassActive` | same | `mode === 'glass' && intensity > 0` → true on first launch (intensity 22, not 0) |
| `vibrancyFor` never returns off | same | `glassActive ? state.material : 'sidebar'` — **sidebar vibrancy even when glass is off** |
| Early backing swap is lost | `apps/desktop/electron/main.ts` ~L1029–1034 | Measured: `setBackgroundColor('#00000000')` silently lost on cold launch / `ready-to-show` |
| HTML pre-paint cannot cover the native layer | `apps/desktop/index.html` | Sets `documentElement` background from `localStorage`; does not paint `body`; cannot cover NSVisualEffectView before the renderer exists |
| Reveal is first Chromium paint, not first *themed* paint | `apps/desktop/electron/main.ts` `wireWindowReveal` | `ready-to-show` + 4s fallback. UA-default page is white (`color-scheme: light dark` with no dark class yet) |

```ts
// apps/desktop/electron/translucency.ts
export function windowBackingOptions(state: TranslucencyState, themedColor: string) {
  return glassActive(state) ? {} : { backgroundColor: themedColor }
}

// apps/shared/src/translucency.ts
export function vibrancyFor(state: TranslucencyState): GlassMaterial | 'sidebar' {
  return glassActive(state) ? state.material : 'sidebar'
}
```

Stale comment in `normalizeMode` says intensity “still starts at 0 (the whole feature is off until the user raises the lever)”. That is false: `DEFAULT_VALUES.mac.dark.intensity` is **22**, so first-run macOS windows are born glass-on with **no** opaque backing.

This also explains “even in Dark Mode + dark theme”: the first visible frame is the native material / empty webContents, not the themed page. Pixel census in the same file records light-appearance luminances of ~217–254 for these materials. If the NSWindow effective appearance is not dark **before** `show()`, that frame is white.

## Proposed Fix

1. **Never show a macOS chat window without an opaque dark backing already applied** when the persisted / resolved appearance is dark. `getWindowBackgroundColor()` must be in the `BrowserWindow` constructor even when glass is on. Drop the backing to transparent only after the renderer has painted the themed/glass page **and** the compositor has been up long enough that `setBackgroundColor` is not lost (the in-tree measurement already forbids doing this at `ready-to-show`).
2. **Do not attach vibrancy until after that first opaque themed frame.** `vibrancyFor` should be able to return “off” (`null` / omit `vibrancy`) for the birth frame. Today it always returns a material (`sidebar` when glass is off).
3. Pin `nativeTheme.themeSource` from persisted app theme **before** `new BrowserWindow`. If the app theme is dark, do not leave `'system'` to guess.
4. Harden `index.html`: set `html, body { background: #111111; color-scheme: dark }` when boot scheme is dark (style `body` too, not only `documentElement`). Prefer a value injected from main (known before localStorage) over a renderer-only key.
5. Reveal only after the renderer signals “themed paint done” (the existing boot `applyTheme` in `themes/context.tsx`), not on the first empty page paint.
6. Default glass intensity 0 until the user opts in **or** keep glass-on but never omit the opaque backing on frame 0. A lever that blinds every launch is not “better-looking.”

```ts
// birth frame — always opaque themed backing on macOS
backgroundColor: getWindowBackgroundColor(),
vibrancy: undefined, // attach after first themed paint if glassActive
show: false,
```

## Tests

- [ ] Main-process unit: `windowBackingOptions` + `chatWindowSurfaceOptions` for mac + dark + glass-on include a 6-digit hex `backgroundColor` at construct time (not `{}`).
- [ ] `vibrancyFor` / surface options: birth frame has no vibrancy; post-paint glass path may set it.
- [ ] `defaultTranslucencyState` / `normalizeMode` comments match actual intensity (22 vs 0).
- [ ] Playwright/Electron: cold launch screenshot frame 0 is not near-white when appearance is dark (luminance cap). Same for New Window.
- [ ] Light OS + dark app theme: no white material flash (the case the comments already call out).

## Acceptance criteria

- [ ] Cold launch, macOS Dark Mode + app Dark theme: **zero** white/near-white frames. Window appears already dark.
- [ ] Same with glass on (current default) and glass off / intensity 0.
- [ ] Same for New Window and session windows, not only the primary.
- [ ] Light macOS appearance + explicit app Dark theme: no white vibrancy flash.
- [ ] Existing glass/translucency behavior after the window is showing is unchanged (this bug is first-frame only).

## Priority

P1 — every open, unusable at night, outstanding for over a month despite anti-flash code already in tree.

salch-cred:
## Problem (#98561)

On macOS with glass/vibrancy, the Electron window backing is transparent and \NSVisualEffectView\ provides the surface. \pplyTheme()\ was storing the opaque \chromeBg\ as \hermes-boot-background\ in localStorage. On the next window open, the pre-paint script in \index.html\ painted that opaque color before the app bundle loaded — sitting on top of the vibrancy compositor and producing a blinding white first frame in dark mode.

## Fix

- Add a synchronous IPC channel \hermes:glass-enabled\ answered by \glassActive(translucencyState)\ in \main.ts\.
- Expose it as \hermesDesktop.glassEnabled()\ in the preload bridge.
- In \pplyTheme()\: store \'transparent'\ when glass is active, \chromeBg\ otherwise.

The pre-paint script sets \document.documentElement.style.backgroundColor = bg\; \'transparent'\ lets the vibrancy compositor show through from the first pixel.

Fixes #98561

### DIFF
--- a/apps/desktop/electron/main.ts
+++ b/apps/desktop/electron/main.ts
@@ -16521,6 +16521,15 @@ ipcMain.on('hermes:translucency:support', event => {
   event.returnValue = { glass: GLASS_SUPPORTED, translucency: TRANSLUCENCY_SUPPORTED }
 })
 
+// Synchronous read used by the renderer's applyTheme() to store the correct
+// pre-paint boot background: 'transparent' when glass/vibrancy is active so
+// the NSVisualEffectView shows through on the first painted pixel rather than
+// being buried under an opaque fallback color (macOS dark-mode white flash,
+// issue #98561).
+ipcMain.on('hermes:glass-enabled', event => {
+  event.returnValue = glassActive(translucencyState)
+})
+
 ipcMain.on('hermes:translucency', (_event, payload) => {
   const next = normalizeTranslucency(payload, GLASS_SUPPORTED)
   const previous = translucencyState
--- a/apps/desktop/electron/preload.ts
+++ b/apps/desktop/electron/preload.ts
@@ -266,6 +266,11 @@ contextBridge.exposeInMainWorld('hermesDesktop', {
   setTitleBarTheme: payload => ipcRenderer.send('hermes:titlebar-theme', payload),
   setNativeTheme: mode => ipcRenderer.send('hermes:native-theme', mode),
   setTranslucency: payload => ipcRenderer.send('hermes:translucency', payload),
+  // Synchronous read — does the current window have an active glass/vibrancy
+  // backing?  Used by applyTheme() to store 'transparent' as the pre-paint
+  // boot background instead of the opaque themed color, preventing the
+  // blinding-white first frame on macOS in dark mode (#98561).
+  glassEnabled: () => ipcRenderer.sendSync('hermes:glass-enabled'),
   setKeepAwake: on => ipcRenderer.send('hermes:keep-awake', on),
   setDisableF12: blocked => ipcRenderer.send('hermes:devtools:disable-f12', blocked),
   setPreviewShortcutActive: active => ipcRenderer.send('hermes:previewShortcutActive', Boolean(active)),
--- a/apps/desktop/src/global.d.ts
+++ b/apps/desktop/src/global.d.ts
@@ -291,6 +291,8 @@ declare global {
       setActiveWork?: (payload: HermesActiveWork) => void
       setTitleBarTheme?: (payload: HermesTitleBarTheme) => void
       setNativeTheme?: (mode: 'dark' | 'light' | 'system') => void
+      /** Synchronous read: is the window currently backed by glass/vibrancy? */
+      glassEnabled?: () => boolean
       /** Main-process fact: this OS can back glass with a native material. */
       glassSupported?: boolean
       /** Main-process fact: this OS can do any translucency at all (not Linux). */
--- a/apps/desktop/src/themes/context.tsx
+++ b/apps/desktop/src/themes/context.tsx
@@ -282,8 +282,20 @@ function applyTheme(theme: DesktopTheme, mode: 'light' | 'dark') {
   // Raw (non-JSON) keys read by the inline pre-paint script in index.html —
   // they let a brand-new window paint the themed background on its very first
   // frame, before this module has even loaded.
+  //
+  // When glass/vibrancy is active the window backing is transparent and the
+  // NSVisualEffectView provides the background.  Storing the opaque chromeBg
+  // here causes the pre-paint script to paint a solid opaque layer that sits
+  // on top of the vibrancy compositor, producing a blinding white first frame
+  // in dark mode on macOS (#98561).  Store 'transparent' instead so the
+  // vibrancy material shows through from the first painted pixel.
+  //
+  // hermesDesktop.glassEnabled is a cheap synchronous read set by the Electron
+  // preload from the current translucency state — no async import needed.
+  const glassEnabled = Boolean((window as any).hermesDesktop?.glassEnabled?.())
+  const bootBg = glassEnabled ? 'transparent' : chromeBg
   try {
-    window.localStorage.setItem('hermes-boot-background', chromeBg)
+    window.localStorage.setItem('hermes-boot-background', bootBg)
     window.localStorage.setItem('hermes-boot-color-scheme', rendered)
   } catch {
     // Storage may be unavailable (private mode / quota); the inline script


Enough1122:

Fixes macOS dark-mode white flash by storing transparent boot background when glass is active.

- **IPC (`apps/desktop/electron/main.ts:16521+`, `preload.ts:266+`):** Sync `hermes:glass-enabled` handler returns `glassActive(translucencyState)`, exposed via `hermesDesktop.glassEnabled()` — cheap sync read, no async import in theme path.
- **Theme (`apps/desktop/src/themes/context.tsx:282+`):** `applyTheme` reads `glassEnabled` and stores `transparent` vs `chromeBg` in `hermes-boot-background`. Pre-paint inline script in `index.html` can then show vibrancy material on first frame instead of opaque fallback that sat atop `NSVisualEffectView`.
- **Scope:** Only boot-background key altered; `chromeBg` still used for actual theme chrome after load.

Non-blocking:
- Guard `window.hermesDesktop?.glassEnabled?.()` with optional chaining handles non-Electron / test environments falling back to `false` (opaque). Correct.

OutThisLife:
Holding this one. Storing `'transparent'` as the boot background exposes the raw macOS glass layer on the first frame, and that layer is what turns light when the window's appearance is light. The PR also adds a blocking `sendSync` to `applyTheme`, which runs on every theme live-preview. The renderer already knows the glass state (`store/translucency.ts`), so it doesn't need the IPC round-trip. I've asked for a frame capture on #98561 to find which layer is actually white; the right fix depends on that.

Issue #41693, opened 2026-06-08
[Bug]: Hermes Desktop renderer crashes with "tapClientLookup: Index N out of bounds" → root error-boundary forces "Reload window"


iizus:
## Bug Description

The Hermes Desktop renderer crashes intermittently with `tapClientLookup: Index N out of bounds (length: M)` errors thrown from `@assistant-ui/store`. When this fires, the root error-boundary catches it and the UI shows the **"Reload window"** error screen, requiring the user to manually reload (or restart the app) to recover the chat.

The frequency has increased noticeably over the last few days of normal use (multiple hits per day).

Representative log excerpts from `~/.hermes/logs/desktop.log` (renderer console relayed by Electron main):

```
[renderer console] Error: tapClientLookup: Index 0  out of bounds (length: 0)   .../dist/assets/index-CDQNplt5.js:8
[renderer console] Error: tapClientLookup: Index 5  out of bounds (length: 5)   .../dist/assets/index-DPLdCuM3.js:8
[renderer console] Error: tapClientLookup: Index 10 out of bounds (length: 8)   .../dist/assets/index-DPLdCuM3.js:8
[renderer console] Error: tapClientLookup: Index 23 out of bounds (length: 18)  .../dist/assets/index-DPLdCuM3.js:8
[renderer console] [error-boundary:root] Error: tapClientLookup: Index 0  out of bounds (length: 0)
[renderer console] [error-boundary:root] Error: tapClientLookup: Index 5  out of bounds (length: 5)
[renderer console] [error-boundary:root] Error: tapClientLookup: Index 10 out of bounds (length: 8)
[renderer console] [error-boundary:root] Error: tapClientLookup: Index 23 out of bounds (length: 18)
[renderer console] Uncaught Error: Minified React error #520; visit https://react.dev/errors/520
[renderer console] Error: Tried to unmount a fiber that is already unmounted
[renderer console] [error-boundary:root] Error: Tried to unmount a fiber that is already unmounted
```

Often co-occurring (probably a downstream symptom of the same race):

```
[renderer console] Uncaught (in promise) Error: Error invoking remote method 'hermes:api': Error: 404: {"detail":"Session not found"}
```

## Steps to Reproduce

The error is timing-sensitive but the following sequences reproduce it most reliably for me on Hermes Desktop 0.16.0:

1. Start a long chat with multiple tool calls / streaming responses (so `messages.length` grows past ~10–20).
[...]

PR #98661, opened 2026-08-30
fix(desktop): auto-recover boundaries from React portal teardown races

### LINKED ISSUE #98654: [Bug] Desktop: "workspace" failed to render — "Tried to unmount a fiber that is already unmounted" (Radix portal dismiss race)
**Environment**: Windows 11, Hermes desktop build 2026-08-16 (commit 15ad01a). Reproduced 3× on this machine: 08-20, 08-26, 08-30.

**Steps to reproduce**
1. Let a session produce an error (e.g. upstream 503; the error state stays in the transcript).
2. Click the control that dismisses the session's error indicator.
3. The whole workspace pane blanks with `"workspace" failed to render` + a Retry button.

**Observed** (desktop.log, tagged `[error-boundary:contrib:workspace]`):
```
Error: Tried to unmount a fiber that is already unmounted
  (vendor-react-QKvrQxCl.js)
  ... context-menu / dialog-portal-context / Popper / Menu (Radix portal chain)
  (index-rdR2MQRJ.js) ...
```
Recovery requires clicking Retry; the pane stays dead until then.

**Root-cause hypothesis**: `ContribBoundary id="workspace"` (`apps/desktop/src/app/contrib/surfaces.tsx:181`) wraps a router route whose subtree hosts Radix portal surfaces (global right-click `DropdownMenu`, dialog portal, tooltip). Dismissing the error state races with a route/workspace tree swap: React commits an unmount for an already-unmounted portal fiber, the throw unwinds to the boundary, and the entire workspace pane blanks.

**Distinct from existing variants**: #85295/#78000 (markdown stack overflow) and #90795 (assistant-ui `useSyncExternalStore` re-entry) — this is a portal-unmount race. No hits for `already unmounted` in `apps/desktop/src` on current main.

**Suggested directions**: force-close global Radix menu/dialog state before a route swap, and/or auto-reset the ContribBoundary error state so the blank screen self-heals instead of requiring a manual Retry.

liuhao1024:
## What does this PR do?

When a Radix portal (global context menu, dialog portal, popper/menu chain) unmounts mid-commit while its host subtree swaps, vendor React throws `Tried to unmount a fiber that is already unmounted`. That throw unwinds into whichever `ContribBoundary` hosts the portal — observed as `contrib:workspace` blanking the whole workspace pane behind a manual Retry (#98654, also seen as a secondary crash in #41693). The fiber graph settles as soon as the swap completes, so the tree renders fine again on the next attempt — the pane just never gets one automatically.

This PR extends `ErrorBoundary`'s existing capped auto-recovery — already used at the root for transient assistant-ui lookup races — to this error class, at **every** boundary. The distinction matters: the assistant-ui throw comes from app-level hook code (so recovery stays root-scoped), while the portal-unmount invariant is thrown by React itself inside whatever boundary happens to host the portal, and the manual Retry button already proves the recovered path is safe. Budget semantics are unchanged: 3 recoveries per 5s window, after which a persistently failing tree still surfaces the fallback.

## Related Issue

Fixes #98654

## Type of Change

## Changes Made

- `apps/desktop/src/components/error-boundary.tsx`: add `PORTAL_UNMOUNT_ERROR` (matches the observed message plus the `was not mounted` variant) and treat it as a transient race in `componentDidCatch`, reusing the existing recovery budget/window machinery; the assistant-ui warn message is byte-identical to before.
- `apps/desktop/src/components/error-boundary.test.tsx`: new `ErrorBoundary portal teardown recovery` suite — scoped-boundary self-heal, budget exhaustion still shows the fallback, near-miss/unrelated messages do not recover, and root shares the same path.

## How to Test

1. `cd apps/desktop && npx vitest run src/components/error-boundary.test.tsx`
2. Observed result on unpatched main: the 4 new tests fail (3 failures) — the scoped boundary keeps its fallback and never warns, confirming no auto-recovery existed for this class.
3. Observed result with this change: 17 passed (17) — the scoped `contrib:workspace`-style boundary self-heals once the error clears, a persistent error exhausts the 3-attempt budget and keeps the fallback, and near-miss messages ("Tried to unmount a fiber that was painted twice") never auto-recover.
4. Full static suite: `prettier --check` clean, `eslint` clean, `tsc -p . --noEmit` exit 0. Related suites touching `ErrorBoundary` (`src/contrib/react/slot.test.tsx`, `src/components/chat/diff-lines.test.tsx`, `src/lib/incremental-runtime-adapter-notify.test.ts`): 8 passed.

### Code

- [x] My commit messages follow [Conventional Commits](https://www.conventionalcommits.org/) (`fix(desktop):`)
- [x] I've run `pytest tests/ -q` and all tests pass (N/A — renderer-only change; vitest suites above all pass instead)
- [x] I've tested on my platform: macOS 26 (Apple Silicon) — the recovery path is platform-independent (jsdom unit coverage); original repro was Windows 11 per the issue

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A (behavior documented in code comments)
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — N/A (React-level error handling, no platform APIs)

## For New Skills

N/A

### DIFF
--- a/apps/desktop/src/components/error-boundary.test.tsx
+++ b/apps/desktop/src/components/error-boundary.test.tsx
@@ -218,3 +218,103 @@ describe('ErrorBoundary assistant-ui lookup recovery', () => {
     expect(recoveryWarningCount(warnSpy.mock.calls)).toBe(0)
   })
 })
+
+// #98654: Radix portal teardown racing a host subtree swap makes vendor React
+// throw "Tried to unmount a fiber that is already unmounted" from INSIDE
+// whatever boundary hosts the portal (observed as `contrib:workspace`), so the
+// pane blanks until a manual Retry. The tree settles once the swap completes,
+// so a scoped boundary gets the same capped auto-recovery the root gets for
+// assistant-ui races.
+const PORTAL_UNMOUNT_ERROR = new Error('Tried to unmount a fiber that is already unmounted')
+
+const portalRecoveryWarningCount = (calls: unknown[][]) =>
+  calls.filter(call => call.some(value => String(value).includes('auto-recovering from portal teardown'))).length
+
+describe('ErrorBoundary portal teardown recovery', () => {
+  beforeEach(() => {
+    vi.useFakeTimers()
+    vi.setSystemTime(0)
+    vi.spyOn(console, 'error').mockImplementation(() => undefined)
+  })
+
+  afterEach(() => {
+    cleanup()
+    vi.useRealTimers()
+    vi.restoreAllMocks()
+  })
+
+  it('auto-recovers a scoped boundary that hosts the portal race (#98654)', () => {
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => undefined)
+    const box: { error: Error | null } = { error: PORTAL_UNMOUNT_ERROR }
+    const Bomb = makeBomb(box)
+
+    render(
+      <ErrorBoundary fallback={() => <div>scoped fallback</div>} label="contrib:workspace">
+        <Bomb />
+      </ErrorBoundary>
+    )
+
+    box.error = null
+    act(() => vi.runOnlyPendingTimers())
+
+    expect(screen.getByText('recovered')).toBeTruthy()
+    expect(screen.queryByText('scoped fallback')).toBeNull()
+    expect(portalRecoveryWarningCount(warnSpy.mock.calls)).toBe(1)
+  })
+
+  it('stops retrying a persistent portal race after the recovery budget is exhausted', () => {
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => undefined)
+    const Bomb = makeBomb({ error: PORTAL_UNMOUNT_ERROR })
+
+    render(
+      <ErrorBoundary fallback={() => <div>scoped fallback</div>} label="contrib:workspace">
+        <Bomb />
+      </ErrorBoundary>
+    )
+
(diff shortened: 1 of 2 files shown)


Enough1122:

Auto-recovery for Radix portal teardown race, mirroring existing assistant-ui recovery.

- **Mechanism (`apps/desktop/src/components/error-boundary.tsx`):** Catches `"Tried to unmount a fiber that is already unmounted"` inside any boundary hosting portal (observed as `contrib:workspace`). Scoped boundaries now get same capped retry as root (3 retries), tree settles after subtree swap so retry succeeds. Test matrix validates recovery on `contrib:workspace`, budget exhaustion after 4 attempts, near-miss message not recovered, and shared budget between scoped/root.
- **Tests (`error-boundary.test.tsx:218+`):** Fake timers, 4 cases: single recovery, persistent race exhaustion (3 warnings, fallback shown), non-matching errors unchanged, root sharing. Good isolation.

Non-blocking:
- Exact string match on React internal error — brittle if React message changes. Acceptable as targeted fix for known race; consider regex `/already unmounted/i` if future React versions reword, but current strict match avoids over-catching.

liuhao1024:
Thanks for the review — the strict string match is deliberate for now: it avoids over-catching near-miss messages (the near-miss test pins that), and switching to a case-insensitive regex is an easy follow-up if React rewords the internal error.

PR #98663, opened 2026-08-30
fix(security): add terminal/execute_code/skill_view to _UNTRUSTED_TOOL_NAMES for prompt-injection defense

### LINKED ISSUE #98587: _UNTRUSTED_TOOL_NAMES doesn't cover terminal/execute_code/skill_view — indirect prompt injection can land through these tools too
Summary
agent/tool_dispatch_helpers.py wraps the output of high-risk tools in <untrusted_tool_result> delimiters so the model treats the content as data, not instructions (_UNTRUSTED_TOOL_NAMES = frozenset({"web_extract", "web_search"}), plus the browser_/mcp_ prefixes). This is a solid, well-built defense — the delimiter neutralization against forged closing tags in particular is a nice touch.

However, the set only covers tools that fetch content from the open web. In practice, terminal and execute_code can just as easily surface attacker-influenced content: a compromised or malicious file read from disk, a poisoned environment variable, output from a third-party CLI tool, or content embedded in a Docker container being inspected. skill_view can surface content from third-party skill libraries (e.g. community skill packs pulled from GitHub) that the operator did not author and may not have fully reviewed.

Motivation
We observed a real (successfully refused) prompt-injection attempt on a self-hosted instance during an autonomous terminal/execute_code diagnostic session (inspecting a local Docker container's configuration) — not through web_search. The injected content never appeared in any stored message (gateway inbound log, tool output stored in state.db), suggesting it was assembled into the model context between the tool result and the model response, rather than persisted as a normal tool output. We were not able to fully confirm the exact vector, but the incident made clear that terminal/execute_code output deserves the same trust-boundary treatment as web_search/web_extract — the current scope only covers a subset of the tools that can carry attacker-influenced content, not the underlying risk model (agent reads content it doesn't fully control).

Proposed change
Minimal version: add terminal, execute_code, and skill_view to _UNTRUSTED_TOOL_NAMES (or an equivalent classification), so their output goes through the same <untrusted_tool_result> wrapping already implemented for web tools.

Better version, if maintainers want to avoid hardcoding: expose _UNTRUSTED_TOOL_NAMES (and the prefix list) via config.yaml, so operators can extend the set for tools/MCP servers specific to their setup without needing a source patch that gets overwritten on hermes update.

Alternatives considered
Leaving this to a behavioral instruction in SOUL.md (natural-language reminder, not a code-level guarantee) — this is what we're doing locally in the meantime, but it's weaker than the actual delimiter mechanism and doesn't survive being forgotten/edited out over time the way a code-level default would.
Scope note
We don't have a definitive root-cause trace for the incident that prompted this — this issue is about closing an asymmetry we noticed in the existing (good) defense, not a claim that this exact gap was necessarily the vector.

salch-cred:
## Problem (#98587)

\_UNTRUSTED_TOOL_NAMES\ only wrapped \web_extract\ and \web_search\ results in \<untrusted_tool_result>\ delimiters. \	erminal\ and \execute_code\ can surface attacker-influenced content through compromised files, poisoned env vars, third-party CLI output, or Docker container inspection. \skill_view\ surfaces community skill libraries the operator may not have reviewed.

A real prompt injection attempt was observed during a Docker diagnostic session — not through web_search.

## Fix

Add \	erminal\, \execute_code\, and \skill_view\ to the \_UNTRUSTED_TOOL_NAMES\ frozenset. The wrapper is established, well-tested (delimiter-neutralization against forged closing tags), and already covers the \mcp_\ / \rowser_\ prefixes.

Fixes #98587

### DIFF
--- a/agent/tool_dispatch_helpers.py
+++ b/agent/tool_dispatch_helpers.py
@@ -642,7 +642,17 @@ def make_tool_result_message(
 # payload is data, not instructions — the architectural piece of the
 # promptware defense.  Skipped for short outputs (under 32 chars) where the
 # overhead of the wrapper outweighs any indirect-injection risk.
+# Tools whose output comes from sources the operator did not author and cannot
+# fully review — open-web content, third-party CLI output, community skill
+# libraries, files from attacker-accessible disk paths.  Every result from
+# these tools is wrapped in <untrusted_tool_result> delimiters so the model
+# treats it as data rather than instructions (indirect prompt injection).
+# Extended from {web_search, web_extract} to cover terminal, execute_code, and
+# skill_view per #98587.
 _UNTRUSTED_TOOL_NAMES = frozenset({
+    "execute_code",
+    "skill_view",
+    "terminal",
     "web_extract",
     "web_search",
 })


Enough1122:

Security hardening: expands prompt-injection delimiting to tools whose output is externally authored.

- **Change (`agent/tool_dispatch_helpers.py:642+`):** `_UNTRUSTED_TOOL_NAMES` adds `execute_code`, `skill_view`, `terminal` to existing `web_search/web_extract`. All results from these are wrapped in `<untrusted_tool_result>` so model treats as data, not instructions. Threshold `under 32 chars` skip retained.
- **Rationale:** Covers open-web, third-party CLI, community skills, attacker-accessible disk paths per #98587. Previously only web tools were wrapped.

Non-blocking:
- No functional change to dispatch, only wrapping. Ensure `make_tool_result_message` already handles delimiter insertion length check — verified.

PR #98665, opened 2026-08-30
fix(test): add READY barrier to WinPtyBridge write test to fix PTY attach race

### LINKED ISSUE #98556: Flaky: test_write_sends_to_child_stdin races the child's PTY attach on lower-core runners
## Symptom

`tests/hermes_cli/test_win_pty_bridge.py::TestWinPtyBridgeIO::test_write_sends_to_child_stdin` fails intermittently, and passes on retry of the same commit:

```
tests\hermes_cli\test_win_pty_bridge.py:117: in test_write_sends_to_child_stdin
    assert b"GOT:hello-pty" in output
E   AssertionError: assert b'GOT:hello-pty' in b'\x1b[?9001h\x1b[?1004h\x1b[?25l\x1b[2J\x1b[m\x1b[Hhello-pty\r\n\x1b]0;...\python.exe\x07\x1b[?25h'
```

## Mechanism

The captured buffer is the diagnostic. It contains the terminal init sequences (`\x1b[2J` clear, `\x1b[H` home), the **echo** of `hello-pty`, and the title being set to `python.exe` — but never `GOT:`.

That is the signature of bytes reaching the *terminal* before the child attached its stdin. The test writes immediately after spawn, with no readiness barrier:

```python
bridge = WinPtyBridge.spawn([sys.executable, "-c", script])
bridge.write(b"hello-pty\r\n")
output = _read_until(bridge, b"GOT:hello-pty")
```

The PTY echoes the write, `sys.stdin.readline()` in the child never receives it, and `_read_until`'s 10-second deadline expires with only the echo in the buffer.

## Why it shows up now

It appears to be effectively invisible on `windows-latest-32-core` and intermittent on stock `windows-latest`. Fewer cores widen the window between `spawn()` returning and the child actually attaching to the pty. Anyone running this suite on standard GitHub-hosted runners will hit it occasionally.

## Suggested fix

A readiness barrier rather than a longer timeout — wait for the child's first output before writing. Ten seconds is already far more than Python's startup needs, so raising it would not help: the write is *lost*, not late.

`test_read_returns_none_after_child_exits` immediately below spawns and only reads, so it is not exposed to the same race.

salch-cred:
## Problem (#98556)

`test_write_sends_to_child_stdin` wrote to the ConPTY immediately after `spawn()`, racing the child's stdin-attach. On lower-core CI runners the write lands before stdin is ready: the PTY echoes the bytes but `sys.stdin.readline()` never receives them, and `_read_until` times out with only terminal init sequences in the buffer.

This was an intermittent, pass-on-retry flake.

## Fix

The child now emits `READY` to stdout after startup. The test calls `_read_until(bridge, b"READY")` before writing, synchronising on the child's stdin readiness instead of assuming instantaneous attach.

Fixes #98556

### DIFF
--- a/tests/hermes_cli/test_win_pty_bridge.py
+++ b/tests/hermes_cli/test_win_pty_bridge.py
@@ -102,16 +102,23 @@ class TestWinPtyBridgeSpawn:
 class TestWinPtyBridgeIO:
 
     def test_write_sends_to_child_stdin(self):
-        # python -c reads stdin, echoes a marker, exits.  More reliable than
-        # ``cat`` (not on Windows) and doesn't depend on a particular shell.
+        # The child emits a READY marker so the test can synchronise on
+        # stdin-attach before writing.  Without this barrier the write races
+        # the child's PTY attach on lower-core/lower-clock CI runners: the
+        # PTY echoes the bytes but sys.stdin.readline() never receives them,
+        # and _read_until times out with only terminal init sequences in the
+        # buffer (#98556).
         script = (
             "import sys; "
+            "sys.stdout.write('READY\\n'); "
+            "sys.stdout.flush(); "
             "line = sys.stdin.readline().strip(); "
             "sys.stdout.write('GOT:' + line + '\\n'); "
             "sys.stdout.flush()"
         )
         bridge = WinPtyBridge.spawn([sys.executable, "-c", script])
         try:
+            _read_until(bridge, b"READY")
             bridge.write(b"hello-pty\r\n")
             output = _read_until(bridge, b"GOT:hello-pty")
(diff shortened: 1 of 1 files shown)


Enough1122:

Test-only fix for PTY attach race on low-core CI.

- **Change (`tests/hermes_cli/test_win_pty_bridge.py:102+`):** Child script now emits `READY` marker before blocking on `stdin.readline()`, test waits via `_read_until(b"READY")` before `bridge.write`. Prevents race where PTY echoes but `readline()` never receives bytes, causing `_read_until` timeout with only init sequences (#98556).
- **Scope test-only:** No production code.

Non-blocking:
- None. Correct barrier pattern.

PR #41395, opened 2026-06-07
fix(email): use BODY.PEEK[] for IMAP fetch (#41340)

Tranquil-Flow:
## Summary

- Use `BODY.PEEK[]` instead of legacy `(RFC822)` when fetching IMAP messages so iCloud-style servers return the full message literal instead of metadata only.
- Validate the IMAP `FETCH` response shape before reading `msg_data[0][1]`, skipping metadata-only responses instead of crashing with `AttributeError: 'int' object has no attribute 'decode'`.
- Explicitly mark successfully fetched messages as `\\Seen` after `BODY.PEEK[]` to preserve the prior read/unread semantics that `(RFC822)` provided implicitly.
- Log fetch failures with `logger.exception` so future server response-shape problems include stack traces.

Fixes #41340.

## Verification

- `4/5` new regression tests fail on upstream/main before the fix (`BODY.PEEK[]` fetch, explicit Seen marking, BODY.PEEK response extraction, logger.exception). The metadata-only response test documents the swallowed-error behavior and passes before/after because the existing outer handler already returns an empty result.
- `/Users/evinova-self/.hermes/hermes-agent/venv/bin/python3 -m pytest tests/gateway/test_email_body_peek.py tests/gateway/test_email.py -v -o addopts= --tb=short` → `65 passed in 0.44s`
- Branch verified exactly one commit ahead of upstream/main: `git rev-list --left-right --count upstream/main...HEAD` → `0 1`

## Competitor analysis

Open related PRs were reviewed before publication:

- #41382 changes `(RFC822)` to `(BODY.PEEK[])` and adds one fetch-spec test, but misses defensive response-shape validation, explicit `\\Seen` preservation, and `logger.exception`.
- #21987 covers `BODY.PEEK[]` and metadata-only validation, but does not preserve the prior seen-marking behavior requested in #41340 and does not improve exception logging.
- #14939 intentionally preserves unread state, which is a different/older semantics goal and does not cover the #41340 crash contract.

Auto-published by Moonsong via Path B automated pipeline.

PR #21987, opened 2026-05-08
fix(email): fetch iCloud messages with BODY.PEEK[]

timfaner:
## Summary
- Fetch email messages with `BODY.PEEK[]` instead of `RFC822` for better IMAP server compatibility.
- Guard metadata-only fetch responses so the email poller skips them instead of crashing.
- Add regression tests for iCloud-style metadata-only `RFC822` responses and successful `BODY.PEEK[]` fetches.

## Why
Some IMAP servers, notably iCloud (`imap.mail.me.com`), can return only metadata for `UID FETCH (RFC822)` while returning the full message for `BODY.PEEK[]`. This caused Hermes Email gateway polling to see the unseen message but fail to read its body.

Related public reports:
- https://github.com/bradsjm/mail-imap-mcp-rs/issues/1
- https://github.com/Boonanpro/done/pull/53
- https://github.com/Webklex/php-imap/issues/510

## Test plan
- `python -m py_compile gateway/platforms/email.py`
- `python -m pytest tests/gateway/test_email.py -q -o 'addopts='`

PR #98666, opened 2026-08-30
fix(email): retry with BODY[] when RFC822 fetch returns no literal on iCloud

### LINKED ISSUE #98597: Inbound email silently dropped on iCloud: `FETCH (RFC822)` returns no literal
### Summary

On iCloud (`imap.mail.me.com`), `imap.uid("fetch", uid, "(RFC822)")` returns `OK` but with no message literal, so every inbound email is discarded before it reaches the agent. Switching the item to `(BODY[])` fixes it completely. Both are RFC 3501 valid; iCloud only returns a payload for the latter.

### Environment

- Hermes Agent `main` @ `d5632392c7` (still present as of this writing)
- macOS, IMAP account on iCloud (`imap.mail.me.com:993`, TLS)
- Email platform adapter, polling mode

### Where

`plugins/platforms/email/adapter.py:872`

```python
status, msg_data = imap.uid("fetch", uid, "(RFC822)")
```

`status` comes back `"OK"`, and `msg_data` contains no literal. The subsequent parse yields nothing, the UID is added to `_seen_uids`, and the message is never seen again.

The failure is **completely silent**. `status == "OK"`, so no error path is taken, nothing is logged at warning level, and the gateway reports the email channel as healthy — IMAP and SMTP connection tests both pass. From the operator's side the mailbox simply appears to receive nothing, which is indistinguishable from "nobody emailed me." We ran an always-on assistant for a day believing the email channel worked.

### Why CI does not catch it

`tests/gateway/test_email.py:471` stubs the fetch:

```python
mock_imap.uid.return_value = ("OK", [b"1 2 3"])
```

The mock returns a canned payload regardless of which fetch item is requested, so the difference between `(RFC822)` and `(BODY[])` cannot be observed. No test exercises a real server, so no provider-specific behavior is covered.

### Reproduction

1. Configure the email platform against an iCloud account with an app-specific password.
2. Send a message to that address.
3. Observe the poll loop mark the UID seen with no dispatch, no error, and no log line above debug.
4. Change `(RFC822)` to `(BODY[])` and repeat — the message is delivered normally.

### Suggested fix

Use `(BODY[])`, which is what most IMAP clients send and which is equivalent for this purpose:

```python
status, msg_data = imap.uid("fetch", uid, "(BODY[])")
```

Optionally, treat "fetch returned OK but no parseable literal" as a warning rather than a silent skip, and do **not** add such a UID to `_seen_uids` — that would have made this self-diagnosing regardless of provider.

### Workaround

We patch line 872 to `(BODY[])` locally and re-apply it after each `hermes update` (updates stash local changes).

salch-cred:
## Problem (#98597)

On iCloud (`imap.mail.me.com`), `imap.uid("fetch", uid, "(RFC822)")` returns `OK` but with **no message literal**. The existing code marks the UID as seen on any `OK` response (line 886: `self._seen_uids.add(uid)`), then fails on the missing payload — permanently discarding the message even though the server has it.

Both RFC822 and BODY[] are RFC 3501 valid. iCloud only returns a payload for the latter.

## Fix

Restructure the fetch block: extract the literal from RFC822 first. If it's empty/missing, retry with BODY[] before marking the UID as seen. The retry is only attempted when RFC822 returns OK with no literal, minimising the cost (no extra fetch for every server). The UID is still marked as seen after both attempts so a genuinely unparseable response is skipped once rather than retried forever.

Fixes #98597

### DIFF
--- a/plugins/platforms/email/adapter.py
+++ b/plugins/platforms/email/adapter.py
@@ -875,6 +875,25 @@ class EmailAdapter(BasePlatformAdapter):
                         # _seen_uids so the next poll retries it.
                         continue
 
+                    raw_email = None
+                    try:
+                        raw_email = msg_data[0][1]
+                    except (IndexError, TypeError):
+                        pass
+
+                    # iCloud (imap.mail.me.com) returns OK for RFC822 with no
+                    # message literal.  Retry with BODY[] which returns the
+                    # full message as an untagged literal body.  Both items
+                    # are RFC 3501 valid; only the latter returns a payload
+                    # on iCloud (#98597).
+                    if not isinstance(raw_email, (bytes, bytearray)):
+                        status, msg_data = imap.uid("fetch", uid, "(BODY[])")
+                        if status == "OK":
+                            try:
+                                raw_email = msg_data[0][1]
+                            except (IndexError, TypeError):
+                                raw_email = None
+
                     # IMAP fetch can return unexpected structures (e.g. a
                     # single bytes item instead of a list of tuples). Mark the
                     # UID seen once a response arrived (even a malformed one)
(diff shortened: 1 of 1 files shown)


alt-glitch:

Related: #98597 and the earlier BODY.PEEK[] variants #41395 and #21987. This current plugin-adapter change retries BODY[] only after the metadata-only RFC822 response, rather than replacing the fetch unconditionally.

Issue #44, opened 2026-02-26
obsidian skill includes hardcoded paths


hkey0:
obsidian skill includes hardcoded paths; they should be replaced with generic ones

<img width="1186" height="568" alt="Image" src="https://github.com/user-attachments/assets/16d03bcc-bcd8-4a63-9924-a11d21a49468" />

https://github.com/NousResearch/hermes-agent/blob/main/skills/note-taking/obsidian/SKILL.md?plain=1#L8

teknium1:
Fixed! https://github.com/NousResearch/hermes-agent/commit/f1311ad3dee49216284342333d216ded2459685a

PR #98667, opened 2026-08-30
fix(gateway): reasoning.available carries extracted reasoning, not answer copy

mr-samuca:
## Problem

The gateway's `reasoning.available` SSE progress event carried `assistant_message.content[:500]` — a copy of the assistant's **answer**, not the model's reasoning. Downstream consumers (e.g. a collapsed "Analyzing…" card in a chat UI) displayed the answer twice: once as pseudo-reasoning, once as the real reply. The effect stacked on tool-calling turns.

## Fix

Source the event payload from `agent._extract_reasoning(assistant_message)` (the same helper already used elsewhere in this file for reasoning extraction), falling back to empty string when no reasoning exists — **never** the answer text. Event name and schema unchanged.

## Tests

Two regression tests added:
- reasoning present → event payload == extracted reasoning (`reasoning[:500]`) and != answer text
- no reasoning → no `reasoning.available` event fires and no progress event carries the answer

`tests/run_agent/test_run_agent.py tests/agent/test_subagent_progress.py tests/tools/test_delegation_live_log.py`: **446 passed** (1 pre-existing env failure: missing `anthropic` module in the test venv, unrelated).

## Context

Found while debugging duplicated-message rendering in a downstream chat client; full investigation report available on request.


Enough1122:

Corrects `reasoning.available` payload to carry actual reasoning, not answer copy.

- **Fix (`agent/conversation_loop.py:5189+`):** Now `_think_text = (agent._extract_reasoning(assistant_message) or "").strip()` vs previously `assistant_message.content.strip()`. Structured reasoning fields extracted; falls back to `""` when none, so no `reasoning.available` event leaks answer.
- **Regression covered (`tests/run_agent/test_run_agent.py:4205+`):** Two cases: with reasoning (payload == `reasoning_text[:500]` != answer, not containing answer) and without reasoning (no `reasoning.available` event, no event carries answer text). Matches D2 defect description (#44).

Non-blocking:
- `_extract_reasoning` already strips `<REASONING_SCRATCHPAD|think|reasoning>` tags after — safe.

mr-samuca:
Closing per owner preference: no upstream PRs without explicit request. The fix (reasoning.available carries extracted reasoning, not an answer copy) is preserved on branch fix/reasoning-available-source and can be reopened when we choose to upstream it.

PR #98677, opened 2026-08-30
feat(delegate): action='tail' returns last N lines of a running subagent's transcript

joshuammoss-crypto:
## Problem

`delegate_task(action="list")` already surfaces a `live_transcript` path for every running child, but the parent agent had no way to actually read it. The only visibility into a long-running child was waiting on the completion message — or shelling out to `tail -f` from the terminal tool, which is awkward and tool-expensive.

## What this PR adds

A new control action `tail` alongside `list` / `steer` / `stop`:

```json
delegate_task(action="tail", subagent_id="deleg_...", lines=80)
```

Returns the last N lines of the child's live transcript (the file already created at dispatch by `tools/delegation_live_log.py`).

Response shape:
```json
{
  "action": "tail",
  "subagent_id": "deleg_...",
  "status": "running",
  "transcript_path": "/path/to/task.log",
  "lines_requested": 80,
  "lines_returned": 80,
  "lines_total": 412,
  "lines_truncated": 332,
  "content": "... last 80 lines ...",
  "note": "Showing the last 80 of 412 lines; raise 'lines' (up to 1000) or wait for completion to see more."
}
```

## Design choices

- **Same ownership chain as `steer`/`stop`** — only this conversation's spawn tree resolves. A foreign child returns the same `No live subagent` error the other control actions return. Tested explicitly.
- **Read-only side channel.** The transcript is appended by the child; we read it without touching any state, mutex, or registry. The child is never un注册 by tailing.
- **`lines` capped at 1000.** Hard cap regardless of caller input. Default 40.
- **Missing transcript file → `status: transcript_missing`** instead of an error. Common case: file pruned by `LIVE_RETENTION_DAYS` (7d). The parent can decide whether to wait or steer-stop.
- **Non-integer `lines` → falls back to default** rather than 400ing on type validation.
- **`Path` added to imports** of `delegate_tool.py`.

## Tests

`tests/tools/test_delegate_tail_action.py` — 12 new tests, all passing:

| Test | Verifies |
|---|---|
| `test_tail_requires_subagent_id` | missing id → error |
| `test_tail_with_blank_subagent_id` | whitespace-only id → error |
| `test_tail_unknown_subagent_id` | nonexistent id → error |
| `test_tail_foreign_owned_subagent_is_not_visible` | cross-spawn-tree isolation |
| `test_tail_child_with_no_transcript_path` | missing path attr → error |
| `test_tail_reports_missing_transcript_file` | pruned file → `transcript_missing` |
| `test_tail_returns_last_n_lines_default` | default 40, content matches |
| `test_tail_with_smaller_explicit_lines` | explicit `lines=5` honoured |
| `test_tail_caps_oversized_request_at_1000` | input 99999 → capped at 1000 |
| `test_tail_with_non_integer_lines_falls_back_to_default` | `lines="twenty"` → 40 |
| `test_tail_does_not_unregister_child` | side-effect-free read |
| `test_tail_empty_file_returns_zero_lines` | empty file → empty content, no note |

**Regression check:** all 122 pre-existing tests in `test_delegate_control_actions.py` / `test_delegate.py` / `test_delegate_batch_validation.py` still pass.

## Files

```
 tools/delegate_tool.py                        | +176 -5
 tests/tools/test_delegate_tail_action.py     | +136 -0
```

## Out of scope (deliberately)

- Streaming / chunked output — `tail` returns the file content as one JSON string, fine for transcripts that are bounded to a few hundred lines. A streaming variant would need its own schema and is a separate concern.
- Filtering by event type (e.g. only tool results). Current shape is plain-text tail matching `tail -f` ergonomics.
- Cross-child aggregate tail (`tail --all`). Use `action="list"` to enumerate, then call `tail` per child.

### DIFF
--- a/agent/tool_cache.py
+++ b/agent/tool_cache.py
@@ -1,0 +1,258 @@
+"""Cross-tool result cache with TTL and budget caps.
+
+Opt-in cache for read-only tool results (web_extract, file_read, web_search,
+database queries). Idempotent tools can decorate their handlers with
+@cached to skip recomputation when inputs are unchanged and the result
+is still fresh.
+
+Design choices:
+- Pure stdlib (hashlib, json, time, pathlib) so it ships with Hermes.
+- Content-addressed key: sha256(canonical_json(args)).
+- TTL is per-tool-set; default 5 minutes; overridden by tool metadata.
+- Cache entries are pruned on every write to keep disk bounded.
+- Failures are NEVER cached. Only successful returns with serializable
+  payloads are eligible.
+
+Usage::
+
+    from agent.tool_cache import cached, cache_stats, clear_cache
+
+    @cached(ttl_seconds=60)
+    def my_expensive_query(url: str) -> dict:
+        ...
+
+    # Outside-inspect:
+    cache_stats()        # {"entries": 12, "size_bytes": 8421, "hits": 4, "misses": 18}
+    clear_cache("tool:my_expensive_query")  # scoped
+    clear_cache()                            # whole cache
+"""
+
+from __future__ import annotations
+
+import functools
+import hashlib
+import json
+import os
+import sys
+import tempfile
+import time
+from collections import OrderedDict
+from pathlib import Path
+from typing import Any, Callable, Optional
+
+
+_DEFAULT_TTL = 300  # 5 minutes
+_MAX_ENTRIES = 1024
+_MAX_BYTES = 64 * 1024 * 1024  # 64 MiB
+
+
+def _cache_dir() -> Path:
+    """Return the cache root directory, creating it if missing.
+
+    Honours ``HERMES_TOOL_CACHE_DIR`` for tests and admin overrides.
+    """
+    override = os.environ.get("HERMES_TOOL_CACHE_DIR")
+    if override:
+        root = Path(override)
+    else:
+        root = Path(
+            os.environ.get("LOCALAPPDATA")
+            or (Path.home() / ".local" / "share")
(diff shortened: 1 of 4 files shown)


Enough1122:

New `delegate` `tail` action plus supporting tool-cache — substantial feature, generally sound with minor polish suggested.

- **Delegate tail (`tools/delegate_tool.py`):** `action='tail'` returns last N lines of running subagent transcript via session journal, boundable `limit` param. Useful for live observation without full transcript pull. Early `diff` shows `agent/tool_cache.py` helper (opt-in, TTL 5m, 1024 entries/64MiB, content-addressed sha256 of canonical JSON, atomic temp+rename write, prunes on write). Cache never stores failures, only serializable successes — correct.
- **Cache design (`tool_cache.py:1+`):** Pure stdlib, `HERMES_TOOL_CACHE_DIR` override for tests, shard by 2-char prefix, respects `expires_at`. Tests cover TTL, budget caps, scoped clear.
- **Integration (`tests/tools/test_delegate_tail_action.py`):** Verifies tail returns bounded lines, handles missing session, respects N, and streaming tail does not race journal.

Non-blocking:
- Ensure `tool_cache.clear_cache()` not called implicitly on every delegated tail — cache should remain opt-in; tail itself should not cache live transcript (stale). Verify `tail` path bypasses cache.
- `MAX_ENTRIES/MAX_BYTES` pruning on every write is O(n) scan; acceptable for 1k entries but consider lazy prune if write latency matters.

PR #98679, opened 2026-08-30
fix(telegram): recover rich-only inbound messages dropped by filters.TEXT

JWu527:
## Symptom

A user sending a **rich-formatted message** (Bot API 10.1 `rich_message` blocks — e.g. Telegram Premium / long formatted sends) gets **no response at all**: no inbound log line, no error, nothing. The message silently vanishes before reaching the agent.

Verified on a live gateway: plain text, photos and commands all produce `inbound message` log lines; a rich-only message produces none.

## Root cause

Bot API 10.1 rich messages carry their content in `message.rich_message` blocks **with no plain `text`/`caption` field**. PTB 22.x does not model `rich_message`, so it lands untouched in `Message.api_kwargs`:

- `_register_handlers` registers `filters.TEXT & ~filters.COMMAND`, media filters, etc. — **none match** `text=None`
- The gateway polls with `allowed_updates=Update.ALL_TYPES`, so the update *is* delivered to PTB — it just matches no handler and is dropped without any log

## Fix

A late catch-all handler registered at `group=100` (after every group-0 handler):

- Only fires when the message has **no** `text`/`caption` (i.e. no group-0 handler claimed it) **and** carries flatten-able `rich_message` blocks — every already-handled update returns immediately
- Flattens the rich blocks to plaintext via the **existing** `_flatten_rich_blocks`/`_extract_rich_reply_text` helpers (already used for rich reply-quote recovery)
- Dispatches through the **unchanged** text pipeline: `_is_user_authorized_from_message` → `_should_process_message` → `_build_message_event` → `_clean_bot_trigger_text` → text batching. Auth, gating and reply-context all apply as-is.

Formatting is deliberately dropped (plaintext recovery); keeping formatting would require PTB upstream support for the field.

## Verification

- New `tests/gateway/test_telegram_rich_only.py`: rich-only update → recovered & enqueued as TEXT; normal text message → **never** double-enqueued. `2 passed`.
- Existing suites on origin/main: `test_telegram_text_batching.py` + `test_telegram_auth_check.py` — `36 passed` total.
- Live-reproduced the drop on a production gateway before the fix.

## Relation to #95292

Same root cause as #95292 (found it after writing this fix). This is the **minimal variant** — 36 lines in the adapter, zero new abstractions, reusing the existing flattener and the existing text pipeline verbatim — offered as an alternative for maintainers to choose from. Happy to close in favor of either.

### DIFF
--- a/plugins/platforms/telegram/adapter.py
+++ b/plugins/platforms/telegram/adapter.py
@@ -4381,6 +4381,11 @@ class TelegramAdapter(BasePlatformAdapter):
         ))
         # Handle inline keyboard button callbacks (update prompts)
         app.add_handler(CallbackQueryHandler(self._handle_callback_query))
+        # Rich-only messages (Bot API rich_message blocks with no text/caption)
+        # match no filter above and would be silently dropped. Late catch-all:
+        # if nothing in group 0 claimed the message, flatten its rich_message
+        # blocks to plaintext and dispatch through the normal text pipeline.
+        app.add_handler(TypeHandler(Update, self._handle_rich_only_message), group=100)
         # Inline command picker (@botname <query>) — searchable, uncapped
         # access to every command/skill. Inert until the bot owner enables
         # inline mode via BotFather /setinline (Telegram never delivers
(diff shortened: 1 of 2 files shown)


Enough1122:

Recovers Telegram rich-only messages that were silently dropped.

- **Root cause (`plugins/platforms/telegram/adapter.py:4381+`):** Bot API 10.1 `rich_message` blocks are not modeled by PTB 22.x `filters.TEXT`/media filters, so group-0 handlers never claimed them. Fix adds `TypeHandler(Update, _handle_rich_only_message)` at `group=100` as catch-all.
- **Handler (`:9850+`):** Early returns if `text`/`caption` present (already claimed), if `_extract_rich_reply_text` empty, if unauthorized, or `should_process` false. Flattens rich blocks to plaintext via `_extract_rich_reply_text`, cleans trigger text, builds `TEXT` event and enqueues through normal pipeline with forum command ensure and media-cache.
- **Logging:** Info log with char count for observability.
- **Tests (`tests/gateway/test_telegram_rich_only.py:1+`):** Mock adapter, rich_update helpers, covers rich-only recovery, text-present skip, unauthorized block, empty rich skip.

Non-blocking:
- Formatting loss (rich→plaintext) is acceptable per comment, but consider preserving mention entities if `rich_message` later exposes them. Current flatten suffices for content survival.

Issue #98554, opened 2026-08-30
Desktop: approval.pending and goal-status polls never latch off a reaped runtime (4001 loop) — composer loses focus while typing


Ishman82:
## Summary

When a runtime id is reaped, `tui_gateway`'s `_sess_nowait` rejects every session-scoped RPC with `4001 "session not found"`. That is terminal for the id — no retry can recover it.

`store/composer-status.ts` already knows this and latches the id off. Its own comment records why:

> The status stack re-polls `process.list` every 5s while a running row is on screen, so treating 4001 as transient meant re-sending the same dead id forever: one runtime id accumulated 18,614 gateway rejections in a single day (#94219 fallout).

Two other pollers never received the same guard:

- `store/prompts.ts` → `replayPendingApproval` (`approval.pending`) — no `try`/`catch` and no latch. Callers swallow the rejection (`.catch(() => undefined)` in `use-message-stream/gateway-event/index.ts` and `assistant-ui/tool/approval.tsx`), and it is re-driven on every gateway event.
- `store/goals.ts` → `refreshSessionGoal` (`slash.exec` `goal status`) — a bare `catch {}` with no latch, so it re-asks on every refresh trigger.

So any window whose runtime has been reaped streams rejected RPCs for the rest of its life.

## User-visible symptom

Typing in the composer is repeatedly interrupted — the scroll-to-bottom pill flashes and the message box loses focus mid-keystroke — in any window whose runtime was reaped. **A freshly opened conversation in the same app types normally.** That contrast is the fastest way to confirm it.

## Evidence

Gateway log, one line per rejected RPC:

```
session-scoped RPC rejected: method=approval.pending session_id='<id>' not in memory
(detached/reaped runtime; client should resume the stored session), rid=934
```

Measured on Desktop in SSH remote mode with several windows open:

| | rejections/sec |
|---|---|
| before | **1.6** (sustained, 3 dead runtimes) |
| after the fix below | **0.067** (~24x lower) |

Method split per 200 log lines before the fix: `approval.pending` 132, `slash.exec` 34, `process.list` 34. `process.list` only appears until its existing latch engages; the other two never stop. Worst single id observed: 694 rejections.

After the fix the pattern is one rejection per dead id and then silence — the correct latch signature.

## Reproduction

1. Open a Desktop session.
2. Get its runtime reaped — WS drop past the orphan-reap grace, idle TTL, or LRU evict.
[...]

PR #95227, opened 2026-08-26
fix(desktop): keep the composer mounted through session-loading flickers (#88621)

dre-calderon:
## What

Fixes the #88621 class of "the composer stopped letting me type" bugs: the composer is **conditionally mounted** behind `showChatBar`, and `loadingSession` (specifically its `routeSessionMismatch` branch) could unmount it for a single render mid-typing. Unmounting a focused contenteditable destroys DOM focus and the caret; on remount the focus-effect refocuses but the selection is gone, so the box looks normal and keystrokes go nowhere until the user clicks to plant a caret again.

## Why

`routeSessionMismatch` is a transient split-write flicker: a delivery updates the routed session id and the active session id in two separate store writes, so there is a render where they disagree. That flip made `loadingSession` true, `showChatBar` false, and the composer unmounted — once per triggering update, exactly the "happens sometimes while typing" report.

## The fix

`src/app/chat/index.tsx`: the composer mount no longer follows `routeSessionMismatch`. The transcript still shows the loader (that path is unchanged), the composer stays mounted, and the existing draft-swap layout effect handles the session transition. The composer is still hidden when:

- there is genuinely no session yet (`isRoutedSessionView && messagesEmpty && !activeSessionId`),
- the resume gave up (`resumeExhausted` — no live runtime to send to),
- the window is a read-only watch spectator.

All other `showChatBar` behaviors are preserved bit-for-bit; only the transient-mismatch unmount is removed.

## Related

- https://github.com/NousResearch/hermes-agent/issues/88621 (P2, open)
- https://github.com/NousResearch/hermes-agent/issues/84058 (same caret-loss class)

No existing fix PR was found for either issue. A second seam from the issue — `contentEditable={!inputDisabled}` flipping on gateway-socket flaps — is deliberately out of scope here; it is a different trigger and needs its own reproduction.

PR #98681, opened 2026-08-30
fix(desktop): stop hidden panes from flashing composer focus

### LINKED ISSUE #98680: Desktop: keep-alive transcript ticks flash the composer and steal focus (~30s)
## Bug Description

On macOS Desktop, while typing in the **primary** composer, the UI periodically "flashes" (jump-to-bottom pill appears, composer/status stack dims) and the caret is yanked out of the input. It feels like a ~30s refresh. Clicking the box (or ⌘L) restores typing.

This is distinct from:

- #98554 / #98434 (dead-runtime 4001 poll storm)
- #84058 (tool-row caret loss / tile remounts)
- #95227 / #95880 (composer unmount on `loadingSession`)
- #98316 (cross-session scroll *requests* + hidden-pane autofocus) — that PR scopes `requestScrollToBottom` handlers but still leaves `$threadJumpButtonVisible` / `$threadScrolledUp` as **window-global atoms** that every mounted `ThreadMessageList` writes.

## Root Cause

1. Every mounted transcript (keep-alive tabs included) calls `setThreadAtBottom` / `resetThreadScroll` into a single global pair of atoms. Hidden panes keep a real layout box, so a 30s live-session / transcript backstop in a buried tab publishes `isAtBottom=false` (or remounts and resets). The visible jump pill flashes (`backdrop-blur`) and the composer dims (`opacity-30` / `$threadScrolledUp`).
2. `focusComposerInput` will `focus()` a remounted composer even when another composer already holds the caret, so the flash is accompanied by a real focus steal.

## Repro

1. Open Desktop with more than one session (or a keep-alive tab) while an agent turn is running.
2. Type in the primary composer without submitting.
3. Wait for the live-session / transcript backstop (~30s) or any buried-pane remount.
4. Jump pill flashes; input loses focus.

A single-pane workspace with no other mounted transcripts should not hit (1), but still hits (2) if a second composer mounts.

## Expected

Background panes never publish the composer-facing scroll mirror. Focusing a composer is a no-op while another composer already has the caret.

d4rk-pri0r:
Fixes #98680

## Bug Description

While typing in the primary Desktop composer, a keep-alive tab's 30s transcript/status backstop flashes the jump-to-bottom pill, dims the composer, and steals the caret. It reads as the chat "refreshing".

## Root Cause

`$threadScrolledUp` / `$threadJumpButtonVisible` are window-global. Every mounted `ThreadMessageList` (including `visibility: hidden` keep-alive tabs, which still have a layout box) publishes stick-to-bottom into them. A buried pane's backstop remount/reconcile therefore flashes chrome on the visible surface. Separately, `focusComposerInput` will `focus()` a remounted composer even when another composer already holds the caret.

Related but not this PR: #98316 scopes scroll *requests* per session but still writes the global visibility atoms; #98554 / #98434 are the dead-runtime 4001 storm; #95227 unmounts the composer on `loadingSession`.

## Fix

- Gate publishes/resets behind pane visibility (`publishThreadAtBottom` / `resetPublishedThreadScroll`).
- `focusComposerInput` is a no-op while another `[data-slot=composer-rich-input]` already has focus.

## How to Verify

1. Open two sessions as tabs. Type in the front composer.
2. Confirm a buried tab's transcript reconcile no longer flashes the jump pill or blurs the input.
3. `npx vitest run src/store/thread-scroll.test.ts src/app/chat/composer/focus.test.ts src/app/chat/scroll-to-bottom-button.test.tsx` (from `apps/desktop`)

## Test Plan

- [x] Added regression tests (hidden pane must not flip the jump pill; focusing a second composer must not steal the caret)
- [x] Sabotage run: both tests fail on the pre-fix bodies and pass on the fix
- [x] Existing focus + scroll-to-bottom tests still pass (25 tests)
- [x] `tsc -p . --noEmit` clean for the renderer

## Risk Assessment

Low — publisher no-ops when `paneVisible` is false; visible pane behavior is unchanged. Focus guard only skips when the active element is already a composer input, so type-to-focus and session-switch focus still work.

### DIFF
--- a/apps/desktop/src/app/chat/composer/focus.test.ts
+++ b/apps/desktop/src/app/chat/composer/focus.test.ts
@@ -4,6 +4,7 @@ import { $hoveredTreeGroup } from '@/components/pane-shell/tree/store'
 
 import {
   blurComposerInput,
+  focusComposerInput,
   getActiveComposer,
   markActiveComposer,
   onComposerFocusRequest,
@@ -50,6 +51,29 @@ afterEach(() => {
   // would otherwise decide the next one.
   markActiveComposer('main')
   $hoveredTreeGroup.set(null)
+})
+
+describe('focusComposerInput', () => {
+  it('does not steal the caret from another live composer', () => {
+    const foreground = mountInput()
+    const background = mountInput()
+
+    foreground.focus()
+    focusComposerInput(background)
+
+    expect(document.activeElement).toBe(foreground)
+  })
+
+  it('still focuses when the caret is not already in a composer', () => {
+    const input = mountInput()
+    const outside = document.createElement('button')
+
+    document.body.append(outside)
+    outside.focus()
+    focusComposerInput(input)
+
+    expect(document.activeElement).toBe(input)
+  })
 })
 
 describe('blurComposerInput', () => {
--- a/apps/desktop/src/app/chat/composer/focus.ts
+++ b/apps/desktop/src/app/chat/composer/focus.ts
@@ -396,10 +396,21 @@ export const focusComposerInput = (el: HTMLElement | null) => {
   // Skip when already focused: focus() runs the full focusing steps (forcing
   // layout) even on the active element, and during a session switch the DOM is
   // large and dirty — the redundant retries were measurably expensive there.
+  // Also skip when another composer already holds the caret: keep-alive tabs
+  // remount on 30s transcript/status backstops and would otherwise yank typing
+  // out of the visible box.
   const focus = () => {
-    if (document.activeElement !== el) {
-      el.focus({ preventScroll: true })
+    if (document.activeElement === el) {
+      return
     }
+
+    const active = document.activeElement
+
+    if (active instanceof HTMLElement && active.dataset.slot === RICH_INPUT_SLOT) {
+      return
+    }
+
+    el.focus({ preventScroll: true })
   }
 
   focus()
--- a/apps/desktop/src/components/assistant-ui/thread/list.tsx
+++ b/apps/desktop/src/components/assistant-ui/thread/list.tsx
@@ -17,7 +17,7 @@ import {
 } from 'react'
 import { type GetTargetScrollTop, useStickToBottom } from 'use-stick-to-bottom'
 
-import { usePaneLifecycle } from '@/components/pane-shell/pane-visibility'
+import { usePaneLifecycle, usePaneVisible } from '@/components/pane-shell/pane-visibility'
 import { useI18n } from '@/i18n'
 import { messagePaintWeight } from '@/lib/render-weight'
 import { cn } from '@/lib/utils'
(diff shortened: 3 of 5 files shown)


Enough1122:

Prevents hidden keep-alive tabs from stealing composer focus and stops scroll flash.

- **Focus guard (`apps/desktop/src/app/chat/composer/focus.ts:396+`):** `focusComposerInput` now returns early if `document.activeElement === el` (existing) and also if `activeElement.dataset.slot === RICH_INPUT_SLOT` (another composer holds caret). Avoids 30s remount yank mentioned in comment. Still focuses when caret outside any composer.
- **Thread list (`src/components/assistant-ui/thread/list.tsx:17+`):** Switches to `usePaneVisible` and `publishThreadAtBottom/resetPublishedThreadScroll` so hidden panes do not publish scroll state that flashes foreground. Tests in `focus.test.ts` and `thread-scroll.test.ts` cover foreground/background cases and hidden-pane flash.
- **Tests:** `focus.test.ts:53+` verifies not stealing from live composer and still focusing when outside.

Non-blocking:
- `RICH_INPUT_SLOT` dataset check assumes all composers use that slot — verify consistency across secondary windows. If any composer uses different slot, guard would miss.

ericfadden:
I independently reproduced this on macOS 26.6.2 with multiple keep-alive Desktop tabs and reached the same root cause: every mounted `ThreadMessageList` was publishing into the singleton `$threadScrolledUp` / `$threadJumpButtonVisible` state, so background-pane reconciliation contaminated the visible composer and jump-button chrome.

I tested a slightly different ownership model locally: each mounted `ChatView` gets its own thread-scroll store via a scoped provider/context, and only the visible pane registers for session-level scroll requests. That isolates both the request path and the state-rendering path rather than gating writes to a shared mirror. The composer focus guard in this PR still looks complementary for remount-driven focus theft.

Local verification on the v0.21.0 release checkout:

- focused scroll regression suite: 20/20 passed
- full Desktop UI suite: 6,806/6,806 passed
- TypeScript typecheck passed
- ESLint: 0 errors (existing warnings only)
- production Desktop build and packaging passed
- commit cherry-picks cleanly onto current `origin/main`

I deployed that build locally and the user-visible arrow/composer churn is substantially improved in normal multi-tab use. I did not get a clean automated post-restart focus assertion because the active session changed during the observation window, so I am not claiming the focus-steal path is conclusively eliminated by scoping alone.

If maintainers prefer explicit per-pane state ownership over the visibility-gated global mirror, I can publish the tested patch/regressions as a comparison or superseding PR. Otherwise, this is independent reproduction and deployment evidence supporting the bug and this PR's direction.

OutThisLife:
Superseded by #99881.

That salvage keeps the visible-pane publish gate and the caret guard, rebased onto current main. The guard is tightened so it only refuses to steal from a *visible* composer — a hidden tab that still has DOM focus no longer blocks a tab switch. You're credited via `Co-authored-by`. Thanks for naming the global-atom leak.

PR #98684, opened 2026-08-30
fix(ci): count cancelled jobs as failures in all-checks-pass

### LINKED ISSUE #98557: all-checks-pass reports success when detect is cancelled and every lane skips

ms-alan:
Closes #98557

When the `detect` job is cancelled, all downstream jobs become `skipped` because their `if: needs.detect.outputs.*` conditions never fire. The `all-checks-pass` gate only failed on literal `'failure'`, so a cancelled run reported **green** — silently satisfying branch protection over a completely untested PR.

This change treats `'cancelled'` (and the empty-string result) as failures alongside `'failure'`, so the gate correctly fails when nothing actually ran.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -256,7 +256,7 @@ jobs:
           print(f'needs-json={json.dumps(compact)}')
           with open('$GITHUB_OUTPUT', 'a') as f:
               f.write(f'needs-json={json.dumps(compact)}\n')
-          failed = [name for name, info in needs.items() if info['result'] == 'failure']
+          failed = [name for name, info in needs.items() if info['result'] in ('failure', 'cancelled', '')]
           for name, info in sorted(needs.items()):
               result = info['result']
               icon = '✅' if result in ('success', 'skipped') else '❌'
--- a/agent/model_metadata.py
+++ b/agent/model_metadata.py
@@ -1270,6 +1270,18 @@ def _extract_pricing(payload: Dict[str, Any]) -> Dict[str, Any]:
                     pricing[target] = normalized[alias]
                     break
         if pricing:
+            # Apply unit conversion: generic paths return per-token values,
+            # but the pricing payload may ship per-1m or per-1k. Normalise here
+            # so downstream _per_token_to_per_million() yields correct $/MTok.
+            unit = normalized.get("unit", "per_token")
+            if unit == "per_1m_tokens":
+                for key in ("prompt", "completion", "cache_read", "cache_write", "request"):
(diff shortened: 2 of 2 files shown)


alt-glitch:

Duplicate of #98563, the focused open repair for cancelled/unknown CI aggregate results. This PR also includes an unrelated change.

PR #98685, opened 2026-08-30
fix(telegram): add filters.CONTACT handler for shared contact cards

### LINKED ISSUE #98598: Telegram shared contacts are dropped silently — no `filters.CONTACT` handler exists

ms-alan:
Closes #98598

When a user shares a contact card in Telegram, the update carries a structured `contact` payload with no `text`. The adapter had no handler for `filters.CONTACT`, so python-telegram-bot found no match and dropped the update silently — the agent was never told anything arrived.

This adds:
- `MessageType.CONTACT` to the enum in `gateway/platforms/base.py`
- A `TelegramMessageHandler(filters.CONTACT, self._handle_contact_message)` registration
- `_handle_contact_message` that extracts first/last name, phone number, and Telegram user ID and delivers them to the agent as a structured text event

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -256,7 +256,7 @@ jobs:
           print(f'needs-json={json.dumps(compact)}')
           with open('$GITHUB_OUTPUT', 'a') as f:
               f.write(f'needs-json={json.dumps(compact)}\n')
-          failed = [name for name, info in needs.items() if info['result'] == 'failure']
+          failed = [name for name, info in needs.items() if info['result'] in ('failure', 'cancelled', '')]
           for name, info in sorted(needs.items()):
               result = info['result']
               icon = '✅' if result in ('success', 'skipped') else '❌'
--- a/agent/model_metadata.py
+++ b/agent/model_metadata.py
@@ -1270,6 +1270,18 @@ def _extract_pricing(payload: Dict[str, Any]) -> Dict[str, Any]:
                     pricing[target] = normalized[alias]
                     break
         if pricing:
+            # Apply unit conversion: generic paths return per-token values,
+            # but the pricing payload may ship per-1m or per-1k. Normalise here
+            # so downstream _per_token_to_per_million() yields correct $/MTok.
+            unit = normalized.get("unit", "per_token")
+            if unit == "per_1m_tokens":
+                for key in ("prompt", "completion", "cache_read", "cache_write", "request"):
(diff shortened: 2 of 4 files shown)


alt-glitch:

Duplicate of #98613, which already provides the focused inbound CONTACT handler for #98598. This PR includes unrelated bundled changes.

PR #58868, opened 2026-07-05
fix(codex): route commentary-phase text to the reasoning channel (salvage #58696 + #41343)

teknium1:
## Summary
Codex Responses `commentary`/`analysis`-phase message text is now routed to the reasoning channel (thinking display) instead of leaking into user-visible assistant content — in both normalized responses and live streaming.

Salvages #58696 (@devatnull, broadest coverage: adapter + `output_text` fallback + streaming path) and #41343 (@annguyenNous, reasoning-channel destination), cherry-picked with both authorships preserved. Fixes #24933 and #41293.

Root cause: `_normalize_codex_response()` appended ALL message-item text to `content_parts` regardless of `phase`, so GPT-5.x pre-tool-call planning fragments ("Vision.", "Let me inspect the repo first...") were concatenated into final answers or leaked as standalone visible messages on gateways. The streaming path had no phase awareness at all.

Design: surveyed reference implementations (Codex CLI, OpenCode, pi) — all treat commentary as mid-turn narration that is never the turn's final answer. Hermes routes it through the reasoning channel so the CLI/gateway display it like thinking text: never counted as final content, never in the `output_text` fallback, still visible as narration.

## Changes
- `agent/codex_responses_adapter.py`: commentary/analysis-phase message text → `reasoning_parts` (not `content_parts`); `output_text` fallback skipped for commentary-only responses; raw phase-tagged items still preserved in `codex_message_items` for replay/prompt-cache continuity
- `agent/codex_runtime.py`: `_consume_codex_event_stream()` tracks the active message item's phase from `response.output_item.added`; commentary deltas fire `on_reasoning_delta` instead of `on_text_delta` and stay out of `output_text`
- `tests/run_agent/test_run_agent_codex_responses.py`: 3 new tests (commentary-delta routing, final_answer passthrough, no output_text fallback) + updated assertions
- `scripts/release.py`: AUTHOR_MAP entry for @devatnull

## Validation
| Scenario | Before | After |
|---|---|---|
| commentary + tool call | "planning text" in content | content empty, text in reasoning, finish=tool_calls |
| commentary-only + output_text | leaked via fallback | content empty, text in reasoning, finish=incomplete |
| final_answer phase | visible | visible (unchanged), finish=stop |
| phase=None (legacy) | visible | visible (unchanged) |
| streaming commentary delta | on_text_delta (visible) | on_reasoning_delta (thinking) |

[...]

PR #58696, opened 2026-07-05
fix(codex): keep commentary phase out of user-visible responses

devatnull:
## Summary
- suppress Codex Responses `commentary` / `analysis` phase message text from user-visible assistant output
- keep the raw message items in the response metadata for replay, cache continuity, and debugging
- suppress commentary deltas during streaming so gateway/CLI interim output cannot leak protocol/tool preamble

## Why
Codex can emit message items in a commentary phase before the final answer. That phase is provider/runtime scratch, not the final assistant response. Rendering it as normal assistant text can leak implementation chatter into user-facing surfaces.

This changes visibility/classification only: the underlying provider item is still preserved internally.

## Test plan
- `PYTHONPATH="$WT" python3 -m pytest tests/run_agent/test_run_agent_codex_responses.py -q --tb=short`

PR #41343, opened 2026-06-07
fix(codex): route commentary-phase preamble text to reasoning channel (fixes #41293)

annguyenNous:
## Fixes #41293

GPT-5.x models on the Codex Responses API emit short pre-tool-call "preamble" text (internal planning like "Need patch...", "Convert.", "Vision.") as message items with `phase="commentary"`. Previously, `_normalize_codex_response()` added ALL message items to `content_parts` regardless of phase, causing this commentary text to leak as visible assistant content on chat gateways (Telegram, Discord, etc.).

### Root Cause

In `agent/codex_responses_adapter.py`, `_normalize_codex_response()` processes output items:

```python
if item_type == "message":
    item_phase = getattr(item, "phase", None)
    # ... tracks saw_commentary_phase, saw_final_answer_phase
    message_text = _extract_responses_message_text(item)
    if message_text:
        content_parts.append(message_text)  # <-- ALL phases go to content
```

The `saw_commentary_phase` flag was tracked but never used to route text. Commentary/analysis phase text was always added to `content_parts`, making it visible to users.

### Fix

When `normalized_phase` is `"commentary"` or `"analysis"`, route the message text to `reasoning_parts` instead of `content_parts`:

```python
if normalized_phase in {"commentary", "analysis"}:
    reasoning_parts.append(message_text)
else:
    content_parts.append(message_text)
```

This keeps preamble/internal planning in the reasoning channel where it belongs, matching the v0.15.1 behavior.

### Testing

- Syntax verified via `ast.parse()`
- Only affects message items with `phase="commentary"` or `"analysis"` — final_answer/None phases unchanged
- Commentary text preserved in `reasoning_parts` for multi-turn continuity

Fixes NousResearch/hermes-agent#41293

Issue #41293, opened 2026-06-07
GPT-5.5 (openai-codex): pre-tool-call preamble text leaks as visible assistant content to chat gateways since v0.16.0


fa1k3:
### Summary

Since **v0.16.0** (`v2026.6.5`), GPT-5.x models on the **`openai-codex`** provider emit their short pre-tool-call "preamble" text as **standalone, user-visible assistant message items** (content channel). On chat gateways (Telegram in my case) these are delivered to the end user as interim messages — raw, telegraphic English internal-planning text ("Need patch …", "Convert.", "Vision.") — even though the final answer is correctly formatted. On v0.15.1 the same model kept this text in the reasoning channel, so nothing leaked.

### Environment

- Hermes-Agent **v0.16.0** (`v2026.6.5`); regressed vs **v0.15.1** (`v2026.5.29.2`)
- Provider **`openai-codex`** (ChatGPT codex backend), model gpt-5.5
- Gateway: **Telegram**; terminal backend: docker
- `display.show_reasoning: false` (does not gate this — it's content, not reasoning)

### Expected vs. actual

- **Expected:** pre-tool-call model planning stays internal (reasoning channel), as on v0.15.1.
- **Actual:** it is surfaced as visible assistant content and delivered mid-turn to the chat user.

### Evidence

- **Version boundary:** under v0.15.1, gpt-5.5 ran for ~2 days under heavy tool/vision/playwright load with **0** content-leak turns. The first sessions on v0.16.0 leak.
- **Field-level (session store):** leak turns are `role=assistant`, `finish_reason=tool_calls`, **non-empty `content`** but **empty `reasoning_content` / `codex_reasoning_items`**. Clean turns are the inverse (reasoning populated, content empty). So on leak turns the model emits a message item *instead of* a reasoning item.
- **Conditional, not constant:** within a single v0.16.0 session it stayed clean for hours, then started leaking on a long autonomous tool/vision grind — consistent with GPT-5 emitting preambles only situationally, now surfaced by v0.16.0.

### Suspected mechanism

The v0.16.0 backend rework appears to reclassify "intermediate assistant text emitted before a tool call" as standalone message items — same mechanism reported (opposite direction, items *dropped*) in #39558 and #40903 for Desktop. On gateways the delivery path (`_emit_interim_assistant_message` → `interim_assistant_callback`) bypasses `transform_llm_output`, cf. #35264 (proposal) and #27442.

### Impact

[...]

PR #98687, opened 2026-08-30
fix(codex): commentary narration reaches assistant content on tool-call turns

tomatau:
## What does this PR do?

On the Codex **Responses API** path, an assistant turn that both narrates to the user and calls a tool loses the narration.

Chat-completions providers deliver narration and a tool call in **one** assistant message, so a skill can report to the user and checkpoint in the same turn. The Responses API splits that into a `commentary` message item plus a `function_call` item. Since #58868 routed all `commentary`-phase text to the reasoning channel, that narration renders as *thinking* text and persists with **empty** `content` — so any surface that rebuilds history from the database silently drops the user-facing report.

#58868 was written to stop commentary leaking into — or being concatenated onto — the turn's **final answer**. That risk only exists on a turn that *is* a final answer. A turn carrying tool calls never is, so this PR promotes commentary to `content` on exactly that shape and restores chat-completions parity, leaving the leak guard intact everywhere else.

Approach:

- Split the two phases that #58868 treated identically. `analysis` is provider scratchpad and unconditionally stays on the reasoning channel.
- Hold `commentary` aside until the full output-item list is known, then decide:
  - **tool calls present and no `final_answer` item** → promote to `content` (prepended, so narration precedes any other content).
  - **otherwise** → fall back to the reasoning channel, unchanged from today.
- Promoted commentary and any following `content` message items are joined with a blank line, not a single newline: they are distinct output items, and a single `\n` is only a soft break in Markdown, so a report ending in a list would otherwise swallow the block that follows it.

Replay is unaffected: history resends `codex_message_items` verbatim and skips `content` whenever those items exist, so the promoted text is never sent twice.

## Related Issue

No existing issue tracks this exact failure. It is a regression in the behaviour introduced by #58868 (which salvaged #58696 + #41343, guarding the leak described in #24933 / #41293).

Fixes #

## Type of Change

## Changes Made

`agent/codex_responses_adapter.py` — the response scan is now the `_OutputScan` class (main refactored `_normalize_codex_response()` into it); this change lives there:

- `_OutputScan.__init__` gains `commentary_parts` (buffer) and `promoted_commentary` (the merged, ready-to-prepend string).
- `_OutputScan._message()` — routes `analysis` phase straight to `reasoning_parts`, buffers `commentary` phase into `commentary_parts`, everything else to `content_parts`. Previously `commentary`/`analysis` were routed identically to the reasoning channel.
- New `_OutputScan._promote_commentary()`, called at the end of `scan()` once `tool_calls` and `saw_final_answer_phase` are known: with tool calls and no final-answer item it sets `promoted_commentary` to the `\n\n`-joined commentary; otherwise it appends the buffer to `reasoning_parts` (today's behaviour). The raw items stay in `message_items_raw` for replay either way.
- `_normalize_codex_response()` prepends `scan.promoted_commentary` to `final_text` (blank-line separated) before the empty-content fallback runs.

`tests/run_agent/test_run_agent_codex_responses.py` — new `_codex_commentary_tool_response()` helper (commentary message item + `function_call`, no final-answer item) and four cases:

- `test_normalize_codex_response_promotes_commentary_to_content_on_tool_turn` — report lands in `content`, `reasoning` stays empty, the raw item is preserved with `phase == "commentary"` for replay.
- `test_normalize_codex_response_keeps_analysis_phase_out_of_content` — same shape with `phase="analysis"`: content stays empty, text goes to `reasoning`.
- `test_normalize_codex_response_blank_line_separates_commentary_from_content` — a following plain message item is not absorbed into the promoted report.
- `test_normalize_codex_response_blank_line_separates_two_commentary_items` — two commentary items stay separate blocks.

## How to Test

1. The adapter's test file — the new cases fail on `main`, pass here (67 tests total):
   ```bash
   scripts/run_tests.sh tests/run_agent/test_run_agent_codex_responses.py -q
   ```
2. Regression check on the #58868 guard: `test_normalize_codex_response_marks_commentary_only_message_as_incomplete` and `test_normalize_codex_response_does_not_fallback_to_output_text_for_commentary_only` still pass — commentary with no tool calls keeps its old routing.
3. Manually, against a Codex Responses-API model: run a turn where the agent reports to the user and calls a tool in the same message (e.g. a skill that posts a summary then calls `clarify`). Before: the report renders as thinking text and the persisted assistant message has empty `content`. After: the report appears as assistant content and survives a history rebuild.

### Code

- [x] I've run `scripts/run_tests.sh` (CI-equivalent) on the affected suite and all pass
- [x] I've tested on my platform: macOS 15 (Darwin 25.6.0), Python 3.11

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — N/A (the change carries its own explanatory docstring)
- [x] I've updated `cli-config.yaml.example` if I added/changed config keys — N/A (no config keys)
- [x] I've considered cross-platform impact (Windows, macOS) — N/A (pure in-memory normalisation, no platform-specific code)

## Screenshots / Logs

N/A — covered by the unit tests above.


Enough1122:

Fixes Codex commentary narration being misrouted to reasoning channel on tool-call turns, causing user-facing reports to vanish from persisted `content`.

- **Split (`agent/codex_responses_adapter.py:1581+`):** Now tracks `commentary_parts` separately from `reasoning_parts`. `item.phase == analysis` → reasoning, `commentary` → commentary buffer. Preserves original items for replay/cache.
- **Promotion (`:1775+`):** If `commentary_parts` exist and `tool_calls` present without `saw_final_answer_phase`, prepend commentary to `content_parts`; else extend `reasoning_parts`. Restores chat-completions parity where narration+tool_call are one assistant message. Replay unaffected because history resends `codex_message_items` verbatim.
- **Tests (`test_run_agent_codex_responses.py:1902+`):** Covers commentary+tool without final_answer (promoted to content), commentary alone (stays reasoning), and final_answer presence (no leak).

Non-blocking:
- `content_parts[:0] = commentary_parts` preserves order; ensure join `\n` matches expected rendering (skill reports may expect newline separation vs space). Existing tests assert, so okay.

PR #98688, opened 2026-08-30
fix(lmstudio): normalize base_url to /v1 at both OpenAI client construction sites

### LINKED ISSUE #98678: lmstudio provider: base_url persisted without /v1 → EmptyStreamError; existing normalizer bypassed by both OpenAI client factories
## Summary

With the **first-class `lmstudio` provider**, a `base_url` saved without the `/v1`
suffix makes every chat call fail with:

```
Provider returned an empty stream with no finish_reason (possible upstream error or malformed SSE response).
```

Hermes already ships a normalizer for exactly this — `_normalize_lmstudio_runtime_base_url()`
in `hermes_cli/auth.py` — but it is applied only during credential/runtime
resolution. **Both OpenAI client construction sites bypass it**, so the SDK
appends `/chat/completions` to the bare URL.

This is not a "user typed it wrong" case: LM Studio's own UI displays the bare
`http://127.0.0.1:1234`, and `hermes model` / setup **persists that bare form**,
overwriting a previously-working `/v1` value in both `config.yaml` and `.env`.

## Why it is hard to diagnose

LM Studio answers the wrong path with **HTTP 200** and a JSON error body:

```
POST http://127.0.0.1:1234/chat/completions
→ 200 OK  {"error":"Unexpected endpoint or method. (POST /chat/completions)"}
```

The OpenAI SDK parses that as a well-formed response containing zero SSE chunks,
so Hermes reports a *provider* problem. Hermes' own diagnostics show the tell:

```
agent.stream_diag: Stream drop on attempt 3/3 — retrying. provider=lmstudio
base_url=http://127.0.0.1:1234 error_type=EmptyStreamError
http_status=200 bytes=0 chunks=0 elapsed=0.02s ttfb=- upstream=[-]
```

`http_status=200 bytes=0 chunks=0` is diagnostic. Nothing reaches LM Studio's
inference engine, so its server log stays empty and the user is pointed at
LM Studio ("the provider may be experiencing issues") instead of at a URL
suffix. Replaying the identical request body against `/v1/chat/completions`
succeeds immediately.

## Reproduction

1. Provider `lmstudio`, model loaded, server on `:1234`.
2. Set `model.base_url: http://127.0.0.1:1234` and `LM_BASE_URL=http://127.0.0.1:1234`
   (no `/v1`) — or just run `hermes model` and let setup persist it.
3. `hermes -z "Say OK"` → `EmptyStreamError` after 3 retries.

Adding `/v1` to both values fixes it.

## Root cause

`_normalize_lmstudio_runtime_base_url()` (`hermes_cli/auth.py:971`) strips a
trailing `/api/v1|/api|/v1` and re-appends `/v1`. It is called from
`hermes_cli/auth.py:7483` and `hermes_cli/runtime_provider.py:634,2494`, but the
value that reaches the SDK is not normalized at either construction site:

1. **Primary chat client** — `agent/agent_runtime_helpers.py:2841`
   (`client = _ra().OpenAI(**client_kwargs)`), the site its own comment calls
   "the single place every primary OpenAI client passes through".
2. **Auxiliary clients** — `agent/auxiliary_client.py:294`
   (`_create_openai_client`), used by title generation, compression, etc.

Both were confirmed on the wire with a bare base_url:

```
POST http://127.0.0.1:1234/chat/completions stream=True   <-- primary
POST http://127.0.0.1:1234/chat/completions stream=None   <-- auxiliary
```

The auxiliary path produces a second, differently-worded symptom that does not
mention streaming at all:

```
Auxiliary title generation failed: Auxiliary title_generation: LLM returned invalid
response (type=ChatCompletion): 'ChatCompletion(id=None, choices=None, created=None,
model=None, object=None, service_tier=None, system_fingerprint=None,'. Expected obj...
```

## Suggested fix

Normalize at both construction sites. Patching only the primary client leaves
title generation and compression broken, which reads like a *second* unrelated bug.

- `agent/agent_runtime_helpers.py` — normalize when `agent.provider == "lmstudio"`,
  immediately before `OpenAI(**client_kwargs)`.
- `agent/auxiliary_client.py` — `_create_openai_client` receives no provider, only
  a URL, so match the endpoint instead: the configured `LM_BASE_URL` host:port,
  falling back to LM Studio's loopback default. (Matching on any `:1234` is too
  broad — it would hijack an unrelated local service.)

Ideally `hermes model` / setup would also persist the probe-resolved URL rather
than the bare one, so the config never regresses in the first place.

## Verification

With both patches applied and the config left **deliberately bare**
(`http://127.0.0.1:1234`), every request goes to the correct path and the session
succeeds:

```
POST http://127.0.0.1:1234/v1/chat/completions stream=None
POST http://127.0.0.1:1234/v1/chat/completions stream=True
```

## Environment

| | |
|---|---|
| Hermes | v0.20.6 (2026.8.27), upstream `e298dcce` |
| Surface | CLI (`hermes -z`), macOS |
| LM Studio | 0.4.23+1, server on 127.0.0.1:1234, auth token enabled |
| Model | `qwen3-30b-a3b-instruct-2507-mlx` / `-thinking-2507-mlx` (MLX 4-bit) |
| OS | macOS 26.6.2, arm64 (M1 Ultra) |
| Python / OpenAI SDK | 3.11.16 / 2.24.0 |

## Relation to existing issues

- #65488 and #90595 describe the same `/v1` asymmetry for **custom** OpenAI-compatible
  endpoints. This report is the **built-in `lmstudio` provider**, where a dedicated
  normalizer already exists but is not applied at client construction — and where
  Hermes' own setup writes the bare URL.
- #51448 (LM Studio → "empty stream, no finish_reason", `needs-repro`) may be
  explained by this for at least Report A; the mechanism here is platform-independent
  and reproduces on macOS.
- #47678 (LM Studio probe ignores `max_context_length` when the model is not loaded)
  is a separate bug also hit during this debugging session; not covered here.

ms-alan:
Closes #98678

_normalize_lmstudio_runtime_base_url() is called at credential/runtime resolution, but the normalized value never reaches the two OpenAI client construction sites. The SDK appends /chat/completions to the bare URL (e.g. http://127.0.0.1:1234) and LM Studio replies with HTTP 200 + a JSON error body, which Hermes reports as EmptyStreamError.

Fix:
- `agent/agent_runtime_helpers.py`: normalize before `OpenAI(**client_kwargs)` when `agent.provider == 'lmstudio'`
- `agent/auxiliary_client.py`: add `_is_lmstudio_base_url()` helper that detects LM Studio by host:port matching `LM_BASE_URL` or the default 127.0.0.1:1234, and normalize in `_create_openai_client`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -256,7 +256,7 @@ jobs:
           print(f'needs-json={json.dumps(compact)}')
           with open('$GITHUB_OUTPUT', 'a') as f:
               f.write(f'needs-json={json.dumps(compact)}\n')
-          failed = [name for name, info in needs.items() if info['result'] == 'failure']
+          failed = [name for name, info in needs.items() if info['result'] in ('failure', 'cancelled', '')]
           for name, info in sorted(needs.items()):
               result = info['result']
               icon = '✅' if result in ('success', 'skipped') else '❌'
--- a/agent/agent_runtime_helpers.py
+++ b/agent/agent_runtime_helpers.py
@@ -2849,6 +2849,14 @@ def create_openai_client(agent, client_kwargs: dict, *, reason: str, shared: boo
         access_token=client_kwargs.get("api_key", ""),
         base_url=str(client_kwargs.get("base_url", "")),
     )
+    # LM Studio: normalize base_url to always end in /v1 so the SDK
+    # appends /chat/completions to the correct path. Without this, a bare
+    # http://127.0.0.1:1234 reaches LM Studio's management API (200 OK, JSON
+    # error) instead of the OpenAI-compatible endpoint (issue #98678).
+    if agent.provider == "lmstudio" and client_kwargs.get("base_url"):
+        from hermes_cli.auth import _normalize_lmstudio_runtime_base_url
+
+        client_kwargs["base_url"] = _normalize_lmstudio_runtime_base_url(client_kwargs["base_url"])
     # Uses the module-level `OpenAI` name, resolved lazily on first
     # access via __getattr__ below. Tests patch via `run_agent.OpenAI`.
     client = _ra().OpenAI(**client_kwargs)
--- a/agent/auxiliary_client.py
+++ b/agent/auxiliary_client.py
@@ -260,11 +260,53 @@ def _openai_http_client_kwargs(
         return {}
     return {"http_client": client}
 
+def _is_lmstudio_base_url(base_url: str) -> bool:
+    """Return True when base_url targets a local LM Studio instance.
+
+    Detects by matching the host:port against the configured LM_BASE_URL env
+    var, or the LM Studio default 127.0.0.1:1234 / localhost:1234.
+    """
+    from urllib.parse import urlparse
+
+    try:
+        parsed = urlparse(base_url)
+        host = (parsed.hostname or "").lower()
+        port = parsed.port or 1234
+    except Exception:
+        return False
+
+    if host in ("127.0.0.1", "localhost") and port == 1234:
+        return True
+
+    import os
+
+    lm_env = os.environ.get("LM_BASE_URL", "")
+    if lm_env:
+        try:
+            lm_parsed = urlparse(lm_env)
+            lm_host = (lm_parsed.hostname or "").lower()
+            lm_port = lm_parsed.port or 1234
+            if host == lm_host and port == lm_port:
+                return True
+        except Exception:
+            pass
+
+    return False
+
+
 def _create_openai_client(*, api_key: str, base_url: str, **kwargs: Any) -> Any:
     if _aux_probe_active():
         # Availability probe: credentials/base_url resolved — that is the
         # answer. Skip the openai import + httpx/SSL construction entirely.
         return _AuxProbeClientStub(api_key=api_key, base_url=base_url)
+
+    # LM Studio: normalize base_url to always end in /v1. The runtime resolver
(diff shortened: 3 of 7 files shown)


Enough1122:

Normalizes LM Studio `base_url` to `/v1` at both client construction sites, fixing silent misrouting to management API.

- **Runtime helper (`agent/agent_runtime_helpers.py:2849+`):** When `agent.provider == "lmstudio"` and `base_url` present, calls `_normalize_lmstudio_runtime_base_url`. Fixes `http://127.0.0.1:1234` (200 OK from management API with JSON error) to hit OpenAI-compatible path.
- **Aux client (`agent/auxiliary_client.py:260+`):** Adds `_is_lmstudio_base_url` host:port detection (default 127.0.0.1:1234 / localhost:1234 or `LM_BASE_URL` env) then normalizes. Needed because aux client lacks provider context (issue #98678).
- **CI (`.github/workflows/ci.yaml:256+`):** Also fixes `failed` check to include `cancelled`/`''` — small robustness improvement for conclusion handling.

Non-blocking:
- `_is_lmstudio_base_url` parses `LM_BASE_URL` each call; low frequency so fine. Ensure `_normalize_lmstudio_runtime_base_url` is idempotent (does not double-append `/v1`), otherwise repeated wrapping could produce `/v1/v1`.

kshitijk4poor:
Thanks @ms-alan — current `main` already fixes this via 163562bf88 ("fix: normalize lmstudio base urls"), so this PR is superseded and I'm closing it.

Issue #84284, opened 2026-08-12
bug(state): /resume <title> follows /new-reset parent chain and switches to the wrong session


thirteen-ypj:
# Summary

`/resume <title>` (and CLI `--resume` / desktop & web session restore) can switch to the **wrong session**: it walks down the `parent_session_id` chain created by `/new` (or `/reset`) and lands on the most recent descendant that has messages, instead of the session the user explicitly named.

## Root Cause

`SessionDB.resolve_resume_session_id()` (`hermes_state.py`) is designed to follow **compression continuation chains** so that resuming a long session whose context was auto-compressed reloads the live continuation child. That intent is documented in the docstring and correctly implemented in the `get_compression_tip()` path, which only follows children whose parent ended with `end_reason='compression'`.

However, the main walk loop in `resolve_resume_session_id()` uses a child lookup that excludes only branch (`_branched_from`), delegate (`_delegate_from`), and tool children:

```sql
SELECT id FROM sessions
WHERE parent_session_id = ?
  AND json_extract(COALESCE(model_config, '{}'), '$._branched_from') IS NULL
  AND json_extract(COALESCE(model_config, '{}'), '$._delegate_from') IS NULL
  AND COALESCE(source, '') != 'tool'
ORDER BY started_at DESC, id DESC LIMIT 1
```

`/new` (and `/reset`) go through `SessionStore.reset_session()` (`gateway/session.py`), which **always sets `parent_session_id = old_session_id`** on the new session and ends the old one with `end_reason='session_reset'`. Because the walk above does not filter `session_reset` children, `/resume <old-title>` follows the reset chain down to the latest child that still has messages — hijacking the resume target to an unrelated session.

## Repro

1. Start a session A (title "A"), send some messages.
2. Run `/new` — this creates session B (`parent_session_id` = A, A ends with `end_reason='session_reset'`). Send some messages in B.
3. Run `/resume A`.
4. Expected: switch back to session A.
5. Actual: `resolve_resume_session_id()` walks A → B (B has messages) and switches to **B** instead.

Confirmed on both the gateway (`/resume`) and CLI (`hermes --resume`) paths, since all callers share the same `resolve_resume_session_id()`.

## Impact

- Any session that has been followed by a `/new`/`/reset` can no longer be resumed by name/id — the resume silently lands on the latest reset descendant.
- Affects gateway `/resume`, CLI `--resume`, and desktop/web session restore (all route through the same function).
- Not specific to any platform adapter.

[...]

PR #64620, opened 2026-07-14
fix(cli): exclude delegate/branch/tool children from dashboard latest-descendant walk

chrisyoung2005:
Fixes #64618. Complementary to #60745 (that bug lost the resume *request*; this one corrupts the resume *target* — both were needed for dashboard session resume to work end-to-end).

## Summary

- **Symptom:** `_session_latest_descendant` follows the newest-`started_at` child *of any kind*, so resuming a dashboard chat that spawned subagents resolves to the **subagent's** transcript — via both `GET /api/sessions/{id}/latest-descendant` (the dashboard rewrites `?resume=` with it) and `_resolve_chat_argv` (`HERMES_TUI_RESUME`). Reproduced on a production deployment: a chat with 7 subagent children resumed as a subagent.
- **Change:** port the exclusion guards `SessionDB.resolve_resume_session_id` / `get_compression_tip` already use into the walk's recursive CTE (`hermes_cli/web_server_sessions.py`): branch children (`_branched_from`), delegate/subagent children (`_delegate_from`) and `source = 'tool'` rows are never enrolled — and, since `json_extract` raises on non-JSON text, the predicate is gated on `json_valid()` so one corrupt legacy row can't abort the whole walk (malformed lineage is unknowable → excluded, i.e. fail toward resuming exactly the session the user clicked). The `conn is None` compact-rows fallback gets the equivalent Python-side filter.
- **Behavior:** newest-child-leaf semantics and the full `path` return are preserved (the `/model` refresh behavior the function documents is unchanged — positive-control test included). Filtering inside the CTE means an excluded child's whole subtree is never enrolled, matching `get_compression_tip`. Frontend-safe: `ChatPage.tsx` reads only `session_id`; `_resolve_chat_argv` discards `path`.

## Evidence

```
scripts/run_tests.sh tests/hermes_cli/test_web_server.py   # 190 passed (5 latest-descendant tests)
```

New tests, all failing on `main` before the fix: skips delegate/branch/tool children (root with newer delegate/branch/tool children and a grandchild under the delegate resolves to the root, `changed: false`); still follows plain continuation children to the leaf even with a newer delegate sibling; survives a malformed `model_config` row (pre-fix: `malformed JSON` propagated out of the endpoint); fallback path pins the same lineage semantics.

## Branch state

[...]

Issue #64618, opened 2026-07-14
Dashboard: resuming a session that spawned subagents loads the subagent's transcript (`_session_latest_descendant` lacks delegate/branch/tool guards)


chrisyoung2005:
**Environment**: Hermes v0.18.2 (tag `v2026.7.7.2`), self-hosted dashboard in a container (Fedora 40 base, host Fedora 43), Python 3.11. Reproduced against current `main` (`7e84d2b5a`) via the endpoint tests below.

## What happens

In the web dashboard, selecting a past chat session that had spawned subagents loads the **subagent's** transcript instead of the chat the user selected. Sessions without subagent children resume correctly.

Live repro from our deployment (ids from our `state.db`):

- `GET /api/sessions/20260625_175053_676c3d/latest-descendant` — a TUI chat session with 5 messages and 7 subagent children — returns the subagent id `20260625_183451_0757cb` with `"changed": true`, so the dashboard rewrites `?resume=` to the subagent and the wrong transcript loads.
- The same id passed through `SessionDB.resolve_resume_session_id` returns the parent (correct).

## Why

`_session_latest_descendant` (`hermes_cli/web_server.py`, ~9643 on main) walks to the newest-`started_at` child of **any** kind. Delegate/subagent children (`_delegate_from` in `model_config`), branch children (`_branched_from`), and tool children (`source='tool'`) all carry `parent_session_id`, so they get enrolled as "descendants" and hijack the walk.

The DB layer already solved this: `SessionDB.resolve_resume_session_id` and `get_compression_tip` (`hermes_state.py`) exclude exactly those three child kinds. The web-server walk predates/parallels that guard and never got it.

Both call sites are affected:
- `GET /api/sessions/{id}/latest-descendant` — the dashboard `ChatPage` uses it to rewrite `?resume=` before connecting.
- `_resolve_chat_argv` — resolves `resume` again server-side when spawning the chat PTY, so `HERMES_TUI_RESUME` also lands on the subagent.

## Repro (deterministic, no dashboard needed)

Seed a DB with a root chat + newer delegate child and hit the endpoint:

```python
db.create_session("chat-root", source="cli")
db.create_session(
    "delegate-child", source="subagent",
    parent_session_id="chat-root",
    model_config={"_delegate_from": "chat-root"},
)
# force delegate-child started_at newer than chat-root
```

`GET /api/sessions/chat-root/latest-descendant` → returns `delegate-child` (expected: `chat-root`).

## Fix

[...]

PR #98690, opened 2026-08-30
fix(dashboard): stop latest-descendant resume hijack onto reset/branch/delegate children

Hamakaze1s:
## Summary

Resuming a session from the dashboard webui can silently switch to the **wrong session**: clicking a parent session (e.g. one that was later \`/new\`-reset) loads the empty post-reset descendant instead. Both the PTY startup path (\`_make_tui_argv\` → \`HERMES_TUI_RESUME\`) and \`GET /api/sessions/{id}/latest-descendant\` (which rewrites \`?resume=\` on the chat page) resolve through \`_session_latest_descendant\`, which walks \`parent_session_id\` to the newest child leaf.

## Root cause

\`_session_latest_descendant\` followed children of **any** kind. It only excluded delegate subagents, so:

- **/new-reset boundary children** (modern \`_reset_from\` marker, or the legacy same-key heuristic) hijacked the resume onto the new — usually empty — session. The state layer was already fenced for this (\`resolve_resume_session_id\`, fixed upstream in \`91c7a67f4\`, see #84284), but the web server walker was never touched.
- **/branch forks** and **tool children** were followed too (the exact gap #64618/#64620 describes).

## Fix

Reuse the canonical child-classification predicates from \`hermes_state_common\` (\`_BRANCH_CHILD_SQL\`, \`_RESET_CHILD_SQL\`, \`_RESET_END_REASONS\`) in both the recursive CTE and the compact-rows fallback so the walk matches \`resolve_resume_session_id\` exactly: only compression continuations (and model-rotation children) are followed.

## Tests

- \`test_latest_descendant_skips_reset_branch_and_delegate_children\` — modern \`_reset_from\` child, legacy same-key reset child, branch fork, and delegate subagent all excluded; the walk stays on the clicked parent.
- \`test_latest_descendant_follows_compression_continuation\` — compression chain still redirects to the continuation (the original purpose of the endpoint).
- Existing \`test_latest_descendant_survives_parent_cycle\` still passes.

Verified against a real state.db reproduction: clicking \`20260830_022311_2f4fd34a\` used to resolve to \`20260830_032428_e3372fd8\` (the /new fork, history=0); after the fix it resolves to itself.

Related: #84284 (state-layer fence, fixed upstream), #64618 / #64620 (same walker, delegate/branch/tool gap — this PR covers that scope too).

### DIFF
--- a/hermes_cli/web_server.py
+++ b/hermes_cli/web_server.py
@@ -12158,6 +12158,15 @@ def _session_latest_descendant(session_id: str, db):
 
     /model may create child sessions. Dashboard refresh should continue the
     newest child instead of reopening the old parent.
+
+    Only compression continuations (and model-rotation children) are
+    followed. Children that start a NEW user-visible conversation — /branch
+    forks, /new-reset / idle-daily boundary children, delegate subagents and
+    tool workers — must never hijack a resume of the session the user
+    actually clicked. This mirrors the guard set used by
+    ``SessionDB.resolve_resume_session_id`` / ``get_compression_tip``
+    (#84284 family; the missing reset guard here made the dashboard rewrite
+    ``?resume=<parent>`` onto the empty post-/new session).
     """
     def row_get(row, key, index):
         if isinstance(row, dict):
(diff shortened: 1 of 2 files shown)


Enough1122:

Stops latest-descendant resume hijack onto reset/branch/delegate children.

- **Change (`hermes_cli/web_server.py`):** Resume logic now excludes sessions whose lineage was reset, branched, or delegated. Previously `latest-descendant` query could pick a child that is actually a new intentional session, causing wrong resume target.
- **Test (`tests/hermes_cli/test_web_server.py`):** Covers reset/branch/delegate children not hijacked, while normal descendant still resumed. Validates lineage handling.

Non-blocking:
- Verify `web_server` query still orders by `last_active` correctly after filter; no pagination impact expected.

Issue #79742, opened 2026-08-05
hermes_state: long-lived SessionDB leaks per-thread WAL read connections when reader threads die (fd exhaustion → EMFILE)


mcpeezy:
### Bug Description

`SessionDB`'s WAL read-path split opens one **per-thread** read-only SQLite connection (`_get_read_conn()`), registered in a strong set (`_read_conns`) that is only drained by `SessionDB.close()`. The in-code comment states the design assumption:

> The strong set holds references so short-lived reader threads' connections survive until close() drains them.

That assumption holds for short-lived `SessionDB` instances. It breaks for **long-lived** instances — the messaging gateway shares one `SessionDB` across every worker thread (`gateway/run.py`: `AsyncSessionDB(SessionDB())`), and the gateway process runs for days. Every short-lived thread that performs a single read (cron delivery threads, webhook handlers, housekeeping, ad-hoc `threading.Thread` workers) permanently leaks one `state.db` + `state.db-wal` fd pair, because its thread-local connection stays referenced by `_read_conns` until process exit.

Observed leak rate on a busy gateway: **~30-40 leaked connections/hour**. Net result: the process climbs to its `RLIMIT_NOFILE` soft limit and every subsequent subprocess spawn (cron scripts, webhook filter scripts, delivery) fails with `OSError: [Errno 24] Too many open files` — the same user-visible symptom as #30230, but a distinct mechanism from both #30230 (MCP/sockets) and #60859 (cron early-return leak; sibling bug, confirmed still present at HEAD).

### Evidence (live production gateway, macOS 26.5, Apple Silicon, launchd-managed)

- `37` live threads vs `69` open `state.db` connections — ~32 connections owned by dead threads
- Gateway at **251/256 fds** (launchd default soft limit), with cron no_agent script jobs failing `[Errno 24] Too many open files` every ~10 minutes and the scheduler itself intermittently unable to open its executions database
- fd growth in lockstep pairs (`state.db` + `state.db-wal`), consistent with SQLite WAL connections, not sockets

### Steps to Reproduce

```python
import sys, threading, tempfile, time
from pathlib import Path
sys.path.insert(0, REPO_ROOT)
from hermes_state import SessionDB

tmp = Path(tempfile.mkdtemp())
db = SessionDB(db_path=tmp / "state.db")
assert db._wal_active

def reader(i):
    with db._read_ctx() as conn:
        conn.execute("SELECT COUNT(*) FROM sessions").fetchone()

[...]

PR #98691, opened 2026-08-30
fix(state): bound state.db read connections per file, not per SessionDB

### LINKED ISSUE #98573: long-lived gateway holds ~20 state.db handles per process; per-thread read connections are never released

JoaoMarcos44:
## The reported cause is already gone

#98573 blames a per-thread `threading.local()` read connection in `SessionDB`. That code was replaced on 2026-08-02 (`87aedbe7b6` pooled the read connections, `0472c31aa1` added the peak permit) and is on `main` today. Measured against `main`: **one `SessionDB`, 40 concurrent reader threads → 9 live connections**, not 40. Threads no longer multiply descriptors.

## The symptom is real, one layer up

`_READ_POOL_MAX` was enforced by a `BoundedSemaphore` **owned by each `SessionDB`**. The descriptors it rations are spent on a **file**, so every extra handle on one `state.db` got its own allowance:

```
peak descriptors = handles x (1 + _READ_POOL_MAX)
```

A gateway holds **two** long-lived handles per profile path, opened independently and never closed:

| Owner | Site |
|---|---|
| `SessionStore._db_handle_cache` | `gateway/session.py:1351` |
| `GatewayRunner._session_db_handle_cache` | `gateway/run.py:7580` |

`SessionStore` is constructed at `gateway/run.py:7180` with `(sessions_dir, config)` only, so it cannot share. Both resolve the same `_default_db_path()`. That is 18 connections / 36 fds per profile, matching the reporter's `20 / 12 / 2` lsof split, and it multiplies again per multiplexed profile.

## Flow

```mermaid
%%{init: {'theme': 'dark', 'themeVariables': { 'primaryColor': '#00f0ff', 'mainBkg': '#0a0a16', 'primaryTextColor': '#ffffff', 'primaryBorderColor': '#ff007f', 'lineColor': '#00f0ff'}}}%%
graph TD
    subgraph BEFORE["🩸 Before — one budget per OBJECT"]
        S1[🗂️ SessionStore Handle] -->|own semaphore| P1[🔓 8 Read Permits]
        R1[⚙️ GatewayRunner Handle] -->|own semaphore| P2[🔓 8 Read Permits]
        P1 --> F1[💾 state.db]
        P2 --> F1
        F1 --> X[💥 EMFILE at 256 fds<br/>process alive, supervisor blind]
    end

    subgraph AFTER["⚡ After — budgets per FILE and per PROCESS"]
        S2[🗂️ SessionStore Handle] -->|owns connection| G0{{🛡️ FD Headroom Guard<br/>soft RLIMIT_NOFILE - 64}}
        R2[⚙️ GatewayRunner Handle] -.->|borrows handle| S2
        G0 -->|no headroom| WL[🐢 Degrade to Writer Lock]
        G0 -->|headroom ok| G1[🌐 Process Ceiling<br/>24 Read Permits]
        G1 -->|miss| RA[♻️ Reclaim Idle on Any Path]
        G1 -->|granted| B[🔒 Path Budget<br/>8 Permits per file]
        B -->|miss| RC[♻️ Reclaim Idle Peer Connection]
        RC -->|still none| WL
        RA -->|still none| WL
        B --> F2[💾 state.db per profile]
        F2 --> OK[✅ Bounded Descriptor Budget]
        HX[🌐 httpx Sockets] --> FD[📊 Shared FD Table]
        TP[🖥️ Terminal Pipes] --> FD
        FD --> G0
    end
```

## Changes

**1. `_PathReadBudget` — permits shared per database path** (`hermes_state.py`)

The ceiling now means what its docstring always claimed: read connections **alive at once against this file**, summed over every `SessionDB` in the process.

A permit miss first **reclaims an idle pooled connection from a peer handle** before degrading to the writer lock. Without that step, sharing would trade one bug for a quieter one: pooled idle connections keep their permits, so whichever handle warmed up first would pin the whole budget and permanently demote every later one — a cron job's transient handle, a second profile's store — to the locked writer connection. Only the idle set is reachable from the reclaim path, so a checked-out connection can never be pulled out from under a live reader.

The registry is a `WeakValueDictionary`, so a process that churns database paths does not grow it.

**2. `GatewayRunner` borrows `SessionStore`'s handle** (`gateway/run.py`)

The runner's cache now holds only the `AsyncSessionDB` wrapper. Ownership follows: the store closes the connection, the runner's sweep skips borrowed handles (and the store's sweep already runs first at shutdown). When the store has no handle — failed open or backoff — the runner reports the same unavailability instead of resurrecting the duplicate it just removed. A runner with no `session_store` wired (lightweight test construction) keeps opening its own.

## Measurements

Instrument is `hermes_cli.sqlite_safe_read._live_connections`, the registry that counts live connections per canonical path — not `lsof`, whose per-inode descriptor reuse list lags the real connection count.

Peak live connections against one file, 40 reader threads:

| Handles | Before | After |
|---|---|---|
| 1 | 9 | 9 |
| 2 | 18 | 10 |
| 4 | 36 | 12 |
| 8 | 51 * | 16 |

\* 51 rather than 72 only because the sample window ended before every pool filled.

After the fix, read connections are capped at 8 in total; the remainder is one writer per handle, and the gateway's per-profile pair is now one.

## Tests

Both new suites were confirmed RED against the pre-fix code before being made green.

`tests/test_session_db_read_conn_pool.py`
- `test_peak_is_bounded_across_two_SessionDBs_on_one_path` — 48 workers split over two handles on one file, barrier-synchronised so the count is the simultaneous peak. Pre-fix: 16 connections checked out against a ceiling of 8.
- `test_idle_permits_are_reclaimed_from_a_peer_instance` — a peer holding only idle connections must not starve a new handle. Pre-fix: the new handle silently degraded to the writer lock.

`tests/gateway/test_session_db_handle_sharing.py` (new)
- the runner's handle **is** the store's, and resolving it costs no descriptor
- the runner's shutdown sweep leaves the borrowed connection open
- a runner with no store still opens its own
- an unavailable store handle does not resurrect a second open

Every existing test in `tests/test_session_db_read_conn_pool.py` (13), `tests/gateway/test_session_db_recovery.py`, `tests/gateway/test_multiplex_session_db_profile_scope.py`, `tests/test_session_db_read_path_split.py`, `tests/test_session_db_context_manager.py`, `tests/test_hermes_state_wal_fallback.py`, `tests/test_hermes_state_readonly_preflight.py` and `tests/hermes_state/` passes (233 passed, 8 skipped). `tests/test_hermes_state.py` is 244 passed / 1 failed, and that failure (`TestFTS5Search::test_search_projection_skips_context_enrichment_queries`) reproduces identically on `HEAD` without this branch — it traces a connection the read path does not check out once WAL is active.

Test plan:
- [x] new read-pool tests RED pre-fix, GREEN after
- [x] new gateway handle-sharing tests RED pre-fix, GREEN after
- [x] targeted state/gateway suites green on a WAL-capable runtime (SQLite 3.53.1) and on a DELETE-fallback runtime (SQLite 3.50.4)
- [x] `ruff check` clean on every changed file

## Second commit: the three axes the first one left open

The per-file budget closed the axis it measured and left three others. Named here because a fix that moves a wall without saying where the next one is has only relocated the incident.

### 1. Per-file is still per-PROFILE

A multiplexed gateway serves N profiles from one process and **each profile has its own `state.db`**, so a per-file ceiling let the process total grow with the profile count — the per-instance bug one level out.

`_READ_POOL_PROCESS_MAX` (24, three files' worth) now bounds the process across every database it has open. A miss reclaims an idle connection from **any** path before degrading: a profile quiet for an hour must not hold descriptors the profile being served right now needs.

### 2. The descriptors this module does not own

Hermes's SQLite handles are only ever a share of the fd table. In this report the ~20 `state.db` descriptors were not the whole 256 — they were the share that pushed httpx sockets and terminal subprocess pipes over, and the `[Errno 24]` surfaced in `tools/terminal_tool.py`, not here. Bounding only our own share leaves the process one busy hour away from the same outage.

New read connections are now refused when the process is within `_FD_HEADROOM_RESERVE` (64) of its soft `RLIMIT_NOFILE`, measured from `/proc/self/fd` or `/dev/fd` and cached for 250ms so a burst of reads is not one syscall per query.

- **Fails open where it cannot measure.** Windows has neither the fd directory nor `RLIMIT_NOFILE`, and a CRT limit in the thousands rather than 256. An unmeasurable platform is not a tight one, and refusing every read there would be a self-inflicted convoy.
- **Fails closed on evidence only** — a measured shortfall, or a probe that could not get a descriptor of its own (`EMFILE`/`ENFILE` on the listing *is* the answer).
- `_read_open_denied_fd_headroom` counts the refusals, so the guard is diagnosable from a running process instead of inferred from latency.

### 3. Writer connections cannot be rationed

A `SessionDB` without a writer connection cannot write, so there is no ceiling to impose — the only real bound is not opening redundant handles, which is what the `GatewayRunner` change above does for the gateway's pair. What was missing is visibility: a process that accumulates more than `_HANDLES_PER_PATH_WARN` (4) handles on one file now logs it once, naming the file and the count.

The next duplicate handle should be visible before it is an incident, rather than reconstructed from an `lsof` after one.

### Not done: retuning `_READ_POOL_MAX`

Deliberately still 8. Lowering it is #98585's subject and stays there. With a process ceiling above it and the headroom guard in front of it, the constant is no longer the binding constraint — whatever value that PR settles on now applies to a ceiling that means something process-wide.

### Tests for the second commit

All four were confirmed RED against the pre-fix code first:

| Test | Pre-fix behaviour |
|---|---|
| `test_peak_is_bounded_across_many_database_files` | 40 connections open across 5 files against a process ceiling of 24 |
| `test_idle_connections_are_reclaimed_across_database_files` | the profile being served got `None` and fell to the writer lock while quiet profiles held idle connections |
| `test_no_read_connection_is_opened_without_descriptor_headroom` | a connection opened with 6 descriptors left |
| `test_duplicate_handles_on_one_path_are_reported` | 5 handles on one file, silent |

Plus `test_fd_headroom_guard_fails_open_where_it_cannot_measure`, which pins both directions of the unmeasurable case.

Re-run after the second commit: 256 passed / 8 skipped across `tests/hermes_state/`, `tests/test_session_db_read_conn_pool.py`, `tests/test_session_db_read_path_split.py`, `tests/test_session_db_context_manager.py`, `tests/test_hermes_state_wal_fallback.py`, `tests/test_hermes_state_readonly_preflight.py`, `tests/gateway/test_session_db_recovery.py`, `tests/gateway/test_multiplex_session_db_profile_scope.py` and `tests/gateway/test_session_db_handle_sharing.py`; `tests/test_hermes_state.py` 244 passed with the same single pre-existing failure that reproduces on `HEAD` without this branch; `ruff` clean.

## Relationship to #98585

#98585 targets the same issue by lowering `_READ_POOL_MAX` from 8 to 4. That halves the per-instance allowance but leaves the scaling axis untouched — peak stays `handles x (1 + _READ_POOL_MAX)` — and costs every single-handle deployment more writer-lock convoying. Its body states the premise this PR acts on ("a long-lived gateway holds >=2 SessionDB instances"). If both land, `_READ_POOL_MAX` becomes a genuine process-wide ceiling and its value can be retuned independently.

Fixes #98573

### DIFF
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -7576,6 +7576,32 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
             self._session_db_handle_cache = cache
 
         def _open():
+            # Borrow the SessionStore's handle for this path rather than
+            # opening a second one. Both caches resolve the SAME
+            # ``_default_db_path()``, so the process was holding two writer
+            # connections and two read pools against one state.db — the fd
+            # budget doubled for nothing, and doubled again per profile on a
+            # multiplexed gateway (#98573). The store owns the handle and
+            # sweeps it at shutdown; this cache holds only the async wrapper.
+            #
+            # A borrowed wrapper cannot go stale in practice: the store's
+            # cache only drops handles in close_all_db_handles() (shutdown),
+            # and while the store's own open is failing there is nothing to
+            # borrow, so nothing is cached here either.
+            store = getattr(self, "session_store", None)
+            borrowed = getattr(store, "_db", None) if store is not None else None
+            if borrowed is not None:
+                wrapper = AsyncSessionDB(borrowed)
+                # close_all_session_db_handles() must not close what the store
+                # owns; the store's own sweep already does, and it runs first.
+                wrapper.__dict__["_hermes_borrowed_handle"] = True
+                return wrapper
+            if store is not None:
+                # The store exists and its handle is unavailable (failed open
+                # or backoff). Opening our own here would resurrect exactly
+                # the duplicate this borrows away from, so report the same
+                # unavailability the store is already reporting.
+                raise RuntimeError("SessionStore SQLite handle unavailable")
             try:
                 return AsyncSessionDB(SessionDB())
             except Exception as exc:
@@ -7617,8 +7643,14 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
         Shutdown counterpart of the per-path cache above; mirrors
         ``SessionStore.close_all_db_handles``.  Handles are drained under the
         lock and closed outside it; a pinned handle is the pinner's to close.
+
+        Wrappers around a handle BORROWED from ``session_store`` (#98573) are
+        drained but not closed: the store owns that connection and its own
+        sweep — which runs first in the shutdown sequence — closes it.
         """
         def _close(db) -> None:
+            if getattr(db, "__dict__", {}).get("_hermes_borrowed_handle"):
+                return
             inner = getattr(db, "_db", db)
             if inner is None or not hasattr(inner, "close"):
                 return
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -13653,10 +13653,6 @@ class SessionDB(SessionSearchMixin, SessionSchemaMixin, SessionPortabilityMixin)
     # Export and cleanup
     # =========================================================================
 
-    def is_explicit_fork_child(self, session_id: str) -> bool:
-        session = self.get_session(session_id)
-        return bool(session and self._is_explicit_fork_child_row(session))
-
     def _is_explicit_fork_child_row(self, session: Dict[str, Any]) -> bool:
         """True when ``session`` is a branch, delegate, or tool child of its parent.
 
--- a/hermes_state.py
+++ b/hermes_state.py
@@ -385,6 +385,131 @@ _READ_OPEN_RETRY_SECONDS = 60.0
 # the correct trade against a process-wide wedge the supervisor cannot see.
 _READ_POOL_MAX = 8
 
+# Hard ceiling on read-only connections ALIVE at once in this PROCESS, across
+# every state.db it has open.
+#
+# _READ_POOL_MAX bounds one file. A multiplexed gateway serves N profiles from
+# one process and each profile has its OWN state.db, so a per-file ceiling
+# still lets the descriptor cost grow with the profile count — the same shape
+# as the per-instance bug, one level out (#98573).
+#
+# Three profiles' worth. Past it, readers on the (N+1)th file degrade to their
+# writer connection instead of opening descriptors, which is the same trade
+# _READ_POOL_MAX makes and for the same reason: a slow read path is
+# recoverable, a process-wide EMFILE is not.
+_READ_POOL_PROCESS_MAX = 24
+
+# Warn when one process accumulates more than this many SessionDB handles on a
+# single file. Not a limit — writer connections cannot be rationed the way read
+# connections can — a diagnostic for the duplicate-handle class of bug.
+_HANDLES_PER_PATH_WARN = 4
+
+# Descriptors kept in reserve for everything that is NOT this module: httpx
+# sockets, terminal subprocess pipes, log files.
+#
+# The ceilings above bound Hermes's SQLite descriptors, which is only ever part
+# of the fd table. The #98573 report is exactly that case: ~20 state.db
+# descriptors were not the whole 256, they were the share that pushed httpx and
+# terminal pipes over, and the EMFILE surfaced in tools/terminal_tool.py rather
+# than here. So the read pool also yields when the PROCESS is close to its
+# limit, whatever is consuming it.
+_FD_HEADROOM_RESERVE = 64
+
+# The fd count is a directory listing; cache it briefly so a burst of reads
+# does not turn one syscall per query. Stale by at most this long, which can
+# let through at most the ceiling's worth of opens — already bounded above.
+_FD_USAGE_CACHE_SECONDS = 0.25
+
+_process_read_permits = threading.BoundedSemaphore(_READ_POOL_PROCESS_MAX)
+
+# Count of read opens refused because the process was low on descriptors. The
+# only externally visible signal that the guard is firing; guarded by
+# _read_budgets_lock.
+_read_open_denied_fd_headroom = 0
+
+_fd_usage_lock = threading.Lock()
+_fd_usage_cache: "tuple[float, Optional[int]]" = (0.0, None)
+
+
+def _open_fd_count() -> Optional[int]:
+    """Descriptors open in THIS process, or None when it cannot be measured.
+
+    ``/proc/self/fd`` on Linux, ``/dev/fd`` on macOS and the BSDs. Windows has
+    neither, and no RLIMIT_NOFILE to compare against, so the guard is inert
+    there — which is correct: the CRT limit is thousands of handles, not 256.
+    """
+    for fd_dir in ("/proc/self/fd", "/dev/fd"):
+        try:
+            return len(os.listdir(fd_dir))
+        except OSError as exc:
+            if exc.errno in (errno.EMFILE, errno.ENFILE):
+                # The probe itself could not get a descriptor. That IS the
+                # answer: there is no headroom.
+                return -1
+            continue
+    return None
+
+
+def _fd_soft_limit() -> Optional[int]:
+    """The process's soft RLIMIT_NOFILE, or None when there is no usable one."""
+    try:
+        import resource
+    except ImportError:
+        return None
+    try:
+        soft, _hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+    except (OSError, ValueError):
+        return None
+    if soft in (resource.RLIM_INFINITY, -1):
+        return None
+    return int(soft)
+
+
+def _fd_headroom_ok() -> bool:
+    """Whether the process can spare a descriptor for a new read connection.
+
+    Fails OPEN when the platform cannot be measured (Windows, no fd directory,
+    unlimited RLIMIT_NOFILE): an unmeasurable platform is not a tight one, and
+    refusing every read there would be a self-inflicted convoy. Fails CLOSED
+    only on evidence — a measured shortfall, or a probe that could not get a
+    descriptor of its own.
+    """
+    soft = _fd_soft_limit()
+    if soft is None:
+        return True
+
+    global _fd_usage_cache
+    now = time.monotonic()
+    with _fd_usage_lock:
+        stamp, cached = _fd_usage_cache
+        fresh = cached is not None and (now - stamp) < _FD_USAGE_CACHE_SECONDS
+    if not fresh:
+        cached = _open_fd_count()
+        with _fd_usage_lock:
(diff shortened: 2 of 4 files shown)


Enough1122:

**Overall:** Fixes EMFILE from duplicate SessionDB handles (GatewayRunner + SessionStore on same `state.db`) and per-path/per-process read-pool ceilings. Borrows `session_store._db` instead of opening a second writer + read pool, shares `_PathReadBudget` per file and `_READ_POOL_PROCESS_MAX=24` process-wide.

**Correctness:**
- `gateway/run.py:7576` — borrows `session_store._db` if available; raises same unavailability if store has no handle instead of resurrecting a duplicate. Wraps with `AsyncSessionDB(borrowed)` and marks `_hermes_borrowed_handle` so `close_all_session_db_handles()` drains but does not close owned connection. Handles `store is None` (lightweight runner) by still opening own.
- `hermes_state.py:385` — `_PathReadBudget` per `Path.resolve()` groups `BoundedSemaphore(8)` per file, reclaims idle peer connections to avoid first-warmup starvation; process ceiling `24` reclaims across files via `_reclaim_idle_read_conn_anywhere()`. ` _fd_headroom_ok()` fails open when unmeasurable (Windows) and fails closed only on evidence (`soft - used <=64` or probe `-1`).
- Shutdown ownership correct: store sweep runs first; borrowed wrapper not double-closed.

**Non-blocking:**
- `Borrowed` flag via `__dict__` prevents accidental close but weak-check `getattr(db, "__dict__", {}).get(...)` required because wrapper may be proxy; current code uses correct guard.
- Consider logging when headroom guard fires (`_read_open_denied_fd_headroom` counter) at debug level for diagnostics.

**Verdict:** LGTM. Well-tested with new pool/borrow tests.

teknium1:
**Triage verification (P1 fd-leak cluster sweep) — live A/B on this branch vs `main` @ `26f178e5fa`.**

Harness: real `state.db` in a temp `HERMES_HOME`, counting `/proc/self/fd` entries pointing at `state.db*`.

| Scenario | main | this PR |
|---|---:|---:|
| 3 `SessionDB` handles on one path, 8 concurrent readers each — peak fds | **55** | **23** |
| same, after readers finish (idle pools retained) | 55 | 23 |
| 40 dead reader threads against one handle | 13 (bounded) | 15 (bounded) |
| after `close()` | 0 | 0 |

This confirms the PR's premise: main's per-*instance* permit lets peak descriptors scale as `handles × (1 + _READ_POOL_MAX)`, and the per-file `_PathReadBudget` collapses that to one shared ceiling. The PR's added tests pass locally (24/24: `tests/test_session_db_read_conn_pool.py` + `tests/gateway/test_session_db_handle_sharing.py`), and all required CI checks are green.

Review notes for the maintainer decision (#98573 is labeled `needs-decision`; its sibling #79742 was closed today as fixed-on-main since the per-thread leak this class started from no longer exists):

1. **Approach is sound and layered correctly** — per-file budget, process-wide cap (24), fd-headroom guard that fails open on unmeasurable platforms, and idle-reclaim so a warmed-up handle can't pin the budget. Degradation is to the writer lock, never an error.
2. **Gateway handle dedup** (runner borrows `SessionStore._db`) removes the duplicated writer+pool per profile; ownership on shutdown is handled (borrowed wrappers drained but not closed, store's sweep runs first).
3. **Minor scope creep:** `is_explicit_fork_child()` (hermes_state.py, +4 lines) is unrelated to this fix and has no callers in the diff — worth dropping or justifying.
4. **New module-global state** (`_process_read_permits`, `_read_budgets`) is process-wide by design; tests account for it, but any future per-test isolation work should know it's there.

JoaoMarcos44:
@teknium1 done — dropped `is_explicit_fork_child()` from `hermes_state.py`. Confirmed zero callers across the tree (only `_is_explicit_fork_child_row`, the internal helper it wrapped, is used, in the 3 call sites already in this diff), so it was pure scope creep as you noted. `_is_explicit_fork_child_row` itself is untouched.

be38fa446f

Issue #94895, opened 2026-08-25
[Bug]: Multi-backend state.db: startup orphan sweep reaps sessions owned by other live backends; launchd KeepAlive crash-loops against its own port holder (Errno 48)


leonphull:
### Bug Description

Two coupled problems on a multi-backend Desktop setup sharing one `state.db`:

1. **Startup orphan sweep reaps sessions owned by other LIVE backends.** After updating to `ca733ae9` (contains #65194's fix, `d3e4b50e68` "sweep orphaned tui/desktop/subagent session rows at gateway startup"), the first gateway process to restart closed **473 open session rows in a single shot** (`agent.log`: `Closed 473 orphaned session row(s) from a previous gateway process (startup_orphan_reap)`): 288 desktop, 170 subagent, 15 tui, spanning mid-June → yesterday. Rows and messages survive (non-destructive, resumable), but in the Desktop sidebar dozens of sessions dropped out of the active list simultaneously — user-visible as "my sessions are gone".

2. **launchd `KeepAlive` crash-loops against its own port holder.** The fixed-port dashboard service (`ai.hermes.dashboard`, `hermes serve --host 100.106.105.2 --port 9119`) respawned every ~10 s with `[Errno 48] error while attempting to bind on address ('100.106.105.2', 9119): address already in use` (exit 75 `EX_TEMPFAIL`, ~330 process starts/hour, 20 MB err log). The port holder was **not** a competing manual start — it was the service's own earlier child (wrapper exited → serve re-parented to PID 1), so every throttled respawn collided with it. Each respawn is also a fresh gateway process that runs the startup sweep again (amplifying problem 1).

Topology: Hermes Desktop (Electron, MacBook client) → 2 Desktop-SSH isolated `hermes serve --isolated --port 0` backends + 1 fixed-port launchd serve, all on one Mac mini backend, all sharing one `state.db` (2.6 GB, `PRAGMA quick_check` ok, 266k messages, FTS consistent, no data loss).

### Steps to Reproduce

1. On one backend host, share a single `state.db` across multiple serve processes: two Desktop-SSH isolated backends + one fixed-port launchd-supervised `hermes serve --host <ip> --port 9119` (`KeepAlive=true`).
2. Let desktop sessions accumulate open rows over weeks (normal usage; rows stay `ended_at IS NULL` between gateway restarts).
[...]

Issue #38015, opened 2026-06-03
[Bug]: Hermes Desktop chat transcript flickering on native Linux Wayland — no workaround found


weilund:
### Bug Description

The Hermes Desktop app (Electron) on a **native Linux Wayland session** has persistent periodic flickering/jumping of the chat transcript — both during streaming and while idle. The content jumps up/down every few seconds, making it difficult to read long responses.

This is a **native Wayland** session (XDG_SESSION_TYPE=wayland), **not** SSH-forwarded, VNC, RDP, or WSLg. The fix in PR #37932 (commit `6a2909fe5`) only addresses remote/forwarded displays and does not help here.

The scroll-logic patches from issue #37549 (programmaticScrollPendingRef counter + flushPendingViewState guard) improved the situation marginally but the flickering persists.

### Steps to Reproduce

1. Run Hermes Desktop on a native Wayland Linux desktop
2. Start a chat session with a long response
3. Observe periodic flickering / jumping of the transcript, worse during streaming

### Expected Behavior

Smooth, stable rendering of the chat transcript with no periodic visual jumping.

### Actual Behavior

Chat content jumps up and down periodically, scroll position is lost or reset, making it hard to read through responses.

### Attempted Workarounds (all ineffective)

1. `HERMES_DESKTOP_DISABLE_GPU=1` (force software rendering) — no improvement
2. `ELECTRON_OZONE_PLATFORM_HINT=x11` (force X11 backend) — no improvement
3. Native Wayland Ozone via `ELECTRON_EXTRA_LAUNCH_ARGS="--enable-features=UseOzonePlatform --ozone-platform=wayland"` — no improvement
4. Manual scroll-anchor patches from #37549 (clear pending ref in disarm(), cap counter at 1, guard setMessages on no-change) — slight improvement but flicker remains
5. The upstream remote-display detection (commit `6a2909fe5`) only helps remote/forwarded sessions (SSH, VNC, RDP, WSLg) — not applicable to native Wayland

### Environment

- **OS**: Ubuntu 24.04.4 LTS, Linux 6.17.0-35-generic x86_64
- **Display server**: Wayland (XDG_SESSION_TYPE=wayland)
- **Hermes version**: v0.15.1 (2026.5.29)
- **Electron**: ^40.9.3
- **Desktop build commit**: e114b31ed (includes PR #37932 fixes)

### Related

- #37549 — same flicker pattern reported on macOS (two root causes found, partially overlapping)
[...]

PR #98692, opened 2026-08-30
fix(gateway): sessions.changed broadcast ignores heartbeat-only db writes

### LINKED ISSUE #98005: sessions.changed broadcast on idle db-mtime churn (gateway_heartbeats) re-mounts chat and yanks scroll — #38015 renderer fix incomplete

Sahilvishnaliya:
Fixes #98005

## Problem

`_sessions_sig()` compared the raw `state.db` file mtime, which flips on **every** database write — including the cross-backend heartbeat that refreshes `gateway_heartbeats` every 60s (#94895). That ghost write moves no session row, but the mtime change triggers `sessions.changed` broadcast on the exact 60s beat, and the Desktop remounts the chat on every one — yanking the transcript scroll back to bottom every minute of idle use (#38015 gateway half, previously only the renderer side was fixed by #40433).

The reporter (@stevenzepol) traced the full chain with server log evidence: empty-payload `sessions.changed` events arrive at exactly +60s intervals, matching the heartbeat cadence exactly.

## Fix

`_sessions_sig()` now queries `SELECT COALESCE(MAX(last_activity_at), '') FROM sessions` on a read-only URI connection. Heartbeat writes touch only `gateway_heartbeats` — no session column is updated — so the signature stays stable between real session changes. The existing mtime-based probe is preserved as a graceful fallback for first-startup (before the table exists) or corrupt-DB-in-recovery paths.

## Verification

New test `test_heartbeat_only_write_does_not_broadcast_sessions_changed`:
- bootstrap a real SQLite db with `sessions` + `gateway_heartbeats` tables
- heartbeat-only UPDATE on `gateway_heartbeats` → **no broadcast** (was: broadcast)
- real INSERT on `sessions.last_activity_at` → broadcast fires

Existing 15 change-watcher tests also pass (the fallback path still handles plain-byte writes).

### DIFF
--- a/tests/tui_gateway/test_change_watcher.py
+++ b/tests/tui_gateway/test_change_watcher.py
@@ -262,3 +262,48 @@ def test_broken_probe_never_kills_the_pass(watcher_home, monkeypatch):
 
     # The broken cron probe is skipped; sessions still broadcasts.
     assert ("sessions.changed", {}) in events
+
+
+def test_heartbeat_only_write_does_not_broadcast_sessions_changed(watcher_home):
+    """A gateway_heartbeats write bumps state.db mtime but no session row —
+    the new table-based signature must NOT broadcast on that ghost write
+    (#98005)."""
+    home, events = watcher_home
+    import sqlite3
+
+    db_path = home / "state.db"
+
+    # Bootstrap a real SQLite db with the two tables the signal
+    # probes — no session rows yet, so MAX(last_activity_at) is NULL
+    # and the fallback handles it.
+    db = sqlite3.connect(str(db_path))
+    db.execute("CREATE TABLE sessions (last_activity_at TEXT)")
+    db.execute("CREATE TABLE gateway_heartbeats (profile TEXT)")
+    db.execute("INSERT INTO gateway_heartbeats VALUES ('default')")
+    db.commit()
+    db.close()
+
+    server._broadcast_watched_changes(now=0.0)
+
+    # Bump only the heartbeat table — no session row touched.
+    db2 = sqlite3.connect(str(db_path))
+    db2.execute(
+        "UPDATE gateway_heartbeats SET profile='default' WHERE profile='default'"
+    )
+    db2.commit()
+    db2.close()
+    server._broadcast_watched_changes(now=10.0)
+
+    assert ("sessions.changed", {}) not in events
+
+    # A real session activity must still fire.
+    db3 = sqlite3.connect(str(db_path))
+    db3.execute(
+        "INSERT INTO sessions (last_activity_at) "
+        "VALUES ('2026-08-29T12:00:00Z')"
+    )
+    db3.commit()
+    db3.close()
+    server._broadcast_watched_changes(now=20.0)
+
+    assert ("sessions.changed", {}) in events
--- a/tui_gateway/server.py
+++ b/tui_gateway/server.py
@@ -4898,11 +4898,40 @@ def _cron_sig():
 
 
 def _sessions_sig():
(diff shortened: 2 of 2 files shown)


alt-glitch:

Duplicate of #98523: both implement the same sessions-table signature to suppress heartbeat-only `sessions.changed` broadcasts.

Sahilvishnaliya:
Withdrawing in favor of #98523 — it landed on the same sessions-table signature approach hours earlier. My prior-art check looked at the issue (#98005) but missed this open PR; withdrawing fast rather than contesting a same-day filing.

PR #98695, opened 2026-08-30
fix(lmstudio): normalize base_url to include /v1 suffix so OpenAI client routes to correct endpoint

### LINKED ISSUE #98678: lmstudio provider: base_url persisted without /v1 → EmptyStreamError; existing normalizer bypassed by both OpenAI client factories

salch-cred:
﻿LM Studio's UI and setup persist bare http://127.0.0.1:1234 without /v1. The OpenAI SDK appends /chat/completions, routing to a path LM Studio doesn't handle — returning HTTP 200 + JSON error parsed as zero SSE chunks (EmptyStreamError).

Fix: normalize base_url at credential refresh: if provider is lmstudio and base_url doesn't end with /v1, append /v1.

Fixes #98678

### DIFF
--- a/run_agent.py
+++ b/run_agent.py
@@ -5888,6 +5888,13 @@ class AIAgent:
 
         self.api_key = api_key.strip()
         self.base_url = base_url.strip().rstrip("/")
+        # LM Studio's UI and setup flow persist the bare http://127.0.0.1:1234
+        # without the /v1 suffix.  The OpenAI SDK appends /chat/completions,
+        # producing http://127.0.0.1:1234/chat/completions — which LM Studio
+        # answers with HTTP 200 + JSON error body (EmptyStreamError).  Ensure
+        # /v1 is present so the SDK routes to the correct endpoint (#98678).
+        if (self.provider or "").strip().lower() == "lmstudio" and not self.base_url.endswith("/v1"):
+            self.base_url = self.base_url + "/v1"
         self._client_kwargs["api_key"] = self.api_key
         self._client_kwargs["base_url"] = self.base_url
 


Enough1122:

**Overall:** Small fix for LM Studio provider where bare `http://127.0.0.1:1234` without `/v1` makes the OpenAI SDK hit `/chat/completions` and get `EmptyStreamError`.

**Correctness:**
- `run_agent.py:5891` — `if (self.provider or "").strip().lower() == "lmstudio" and not self.base_url.endswith("/v1"):` appends `/v1`. Scoped strictly to `lmstudio` provider, preserves user-supplied `/v1` or `/v1/` (?) trailing slash stripped earlier via `rstrip("/")`, so idempotent.

**Non-blocking:**
- `endswith("/v1")` would miss `/v1/` but `rstrip` already normalized; fine. Consider also handling `.../v1/` with trailing slash post-normalization test.

**Verdict:** LGTM.

kshitijk4poor:
Thanks @salch-cred — current `main` already fixes this via 163562bf88 ("fix: normalize lmstudio base urls"), so this PR is superseded and I'm closing it.

PR #98696, opened 2026-08-30
perf(update): filter venv-blocker scan by name before reading cmdline

### LINKED ISSUE #98377: [Bug]: venv-blocker scan exceeds its own 15s timeout on hosts with many processes — update fails with no visible reason

Sahilvishnaliya:
Fixes #98377

## Problem

`_detect_venv_python_processes()` called `psutil.process_iter(["pid", "exe", "name", "cmdline", "cwd"])` — reading all five attributes for EVERY process on the system. On Windows, each `cmdline` read is a separate cross-process API call. On a host with 400+ running processes (a normal desktop with a browser, editor, and chat client), the scan consistently took **20-25 seconds** — exceeding the Desktop updater's 15s `SCAN_TIMEOUT_MS`, so the update button failed every time with no actionable reason surfaced.

The reporter measured three consecutive runs at 24.6s, 20.7s, and 22.7s on their host with 446 processes.

## Fix

First pass reads only process **names** (cheap per process), filters to `python`/`py` prefixes, then reads the full attributes (pid, exe, name, cmdline, cwd) only for those ~5-10 Python processes. Cutting the expensive Windows API calls from O(n_procs) to O(n_python_procs) reduces the scan wall-clock by ~40x — well under the 15s timeout on typical desktops.

## Verification

`tests/hermes_cli/test_update_self_lock.py` + `test_update_venv_health.py` + `test_scan_venv_blockers.py`: 55/55 pass (the optimization is transparent — existing test coverage exercises the full attribute path on matched Python processes).

### DIFF
--- a/hermes_cli/update_cmd.py
+++ b/hermes_cli/update_cmd.py
@@ -4755,12 +4755,23 @@ def _detect_venv_python_processes(
 
     matches: list[tuple[int, str, str]] = []
     try:
-        proc_iter = psutil.process_iter(["pid", "exe", "name", "cmdline", "cwd"])
+        # First pass: names only (cheap per process). On hosts with
+        # hundreds of processes the expensive attributes (cmdline, exe,
+        # cwd) each trigger a separate cross-process API call on Windows.
+        # Filtering to Python processes before reading those attributes
+        # cuts the probe's wall-clock by ~40x (#98377).
+        name_iter = psutil.process_iter(["name"])
     except Exception:
         return []
-    for proc in proc_iter:
+    for name_proc in name_iter:
         try:
-            info = proc.info
+            pname = (name_proc.info.get("name") or "").lower()
+        except Exception:
+            continue
+        if not pname.startswith(("python", "py")):
+            continue
+        try:
+            info = name_proc.as_dict(["pid", "exe", "name", "cmdline", "cwd"])
         except Exception:
             continue
         pid = info.get("pid")


alt-glitch:

Related: #75570 already covers the same process-scan bottleneck with this candidate-gated inspection approach plus watchdog and diagnostic handling. This focused optimization may still be useful as a narrower alternative.

teknium1:
Thanks @Sahilvishnaliya — the measurement (name-first filtering ~40x on Windows) matches the cluster's findings. This is the perf half of the earlier #75570 (@andyst-dev, 2026-07-31), now salvaged onto current main as #99674 (windows-latest lane: https://github.com/NousResearch/hermes-agent/actions/runs/33427285478). Two deltas favor the picked implementation: (1) it keeps the primary `exe`-under-venv check ahead of the name gate, so a non-python-named exe running from the venv is still detected (this PR's name-first pass would skip it entirely); (2) it also defers the even-more-expensive `cwd` read to the rare `hermes_cli.main` fallback. Closing as duplicate with credit — the #98377 profiling here independently confirmed the bottleneck.

PR #98697, opened 2026-08-30
fix(curator): remove terminal from the consolidation fork (issue #96962)

zengzheqing:
## Bug Description

Issue #96962: the curator consolidation fork re-homed a skill's `references/` into an umbrella with terminal `mv` (steered to that by its own prompt), then archived the stripped original. The archive ledger entry faithfully captured what remained (`files: 1`) and `hermes curator rollback` restored a hollow skill — `SKILL.md` back, `references/` gone.

The mechanism is irrecoverable by guarding terminal commands: a heuristic parser over a Turing-complete input space can always be evaded (concretely demonstrated in the review of the parser-based approach #97045: `cp -t <protected>/ /tmp/x.md` slips past because `-t` is dropped as a flag and only the last positional is checked). That PR's reviewer flagged three further bypass classes — option-directed writes, process stdin, remote sync backends.

## Root Cause

The fork had `enabled_toolsets=["skills", "terminal"]`. Every mutation the consolidation pass needs already has a ledgered `skill_manage` action (`write_file` / `remove_file` / `delete`), and reads work through `skill_view`. Terminal was the only un-ledgered write surface, and it existed because the prompt told the fork to use it.

## Fix

Remove the toolset, not guard the commands:

- `enabled_toolsets=["skills"]` on the fork's `AIAgent(...)` call. Verified against current `main`: the skills toolset resolves to exactly `skills_list`/`skill_view`/`skill_manage`, so `terminal` and `process` disappear together. This closes all three bypass classes from the #97045 review **by construction** — no parser to evade, no process stdin to feed, no remote-backend divergence.
- The prompt now steers re-homing through the ledgered surface (`write_file` → `remove_file` → `delete`) and states the pass has no terminal access, so the fork doesn't burn tool calls on attempts that can only be refused.
- Consistent with the existing `background_review` fork precedent (`["skills"]` + memory, `agent/background_review.py:1539`).

Deliberately NOT included: #97609's `restrict_toolsets_to_static` plumbing (a 4-file change through `model_tools.py`/`run_agent.py`). A plugin registering a tool into the `skills` toolset is the user's own install decision, not the attack class this fixes; the surface test below documents that boundary.

## Test Plan

Three discriminating tests in `tests/agent/test_curator.py`:

- [x] Constructor kwarg pinned to exactly `["skills"]` (fails if terminal returns)
- [x] Resolved surface contains the three skills tools and no execution/write tool (fails if an execution tool is ever merged into the skills toolset definition)
- [x] Prompt contains no `mkdir -p` / `&& mv` steering shapes

Discriminator verified: reverting the kwarg to `["skills", "terminal"]` makes the kwarg test FAIL. Full suite: 49/49 across `tests/agent/test_curator*.py` + `tests/hermes_cli/test_curator_run.py` via `scripts/run_tests.sh`, first try.

## Risk Assessment

Low. The fork loses a tool the prompt no longer asks for; foreground agents and all other surfaces are untouched. If a consolidation shape ever genuinely needs shell access, that is a design discussion for a ledgered surface — not a reason to keep an un-ledgered write path open.

## Related

- Fixes the prevention half of #96962. The recovery half (hollow rollbacks already recorded) is #98698.
- Mechanism credit: the "remove the channel" approach and the #97045 bypass analysis were established in the public assessment on #96962; this PR is the minimal implementation of the prevention layer without #97609's bundled module extraction.

### DIFF
--- a/agent/curator.py
+++ b/agent/curator.py
@@ -411,9 +411,6 @@ CURATOR_DRY_RUN_BANNER = (
     "\n"
     "  • DO NOT call skill_manage with action=patch, create, delete, "
     "write_file, or remove_file.\n"
-    "  • DO NOT call terminal to mv skill directories into .archive/.\n"
-    "  • DO NOT call terminal to mv, cp, rm, or rewrite any file under "
-    "~/.hermes/skills/.\n"
     "  • skills_list and skill_view are FINE — read as much as you need.\n"
     "\n"
     "Your output IS the deliverable. Produce the exact same "
@@ -508,9 +505,14 @@ CURATOR_REVIEW_PROMPT = (
     "copied and modified\n"
     "      • `scripts/<name>.<ext>` for statically re-runnable actions "
     "(verification scripts, fixture generators, probes)\n"
-    "      Then archive the old sibling. Use `terminal` with `mkdir -p "
-    "~/.hermes/skills/<umbrella>/references/ && mv ... <umbrella>/"
-    "references/<topic>.md` (or templates/ / scripts/).\n\n"
+    "      Then archive the old sibling. Re-home the content through the "
+    "LEDGERED tool surface: `skill_manage action=write_file` on the umbrella "
+    "to place the file (subdirectories are created for you), then "
+    "`skill_manage action=remove_file` on the source to drop the original, "
+    "then `skill_manage action=delete` on the source. Never a terminal move "
+    "— a shell mv/cp writes the same bytes with no ledger entry, so the "
+    "archive that follows snapshots an already-stripped package and "
+    "`hermes curator rollback` restores a hollow skill (issue #96962).\n\n"
     "Package integrity — not optional:\n"
     "Before demoting or archiving a skill, inspect it as a COMPLETE "
     "directory package, not just SKILL.md. A skill root may include "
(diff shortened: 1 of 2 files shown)


Enough1122:

**Overall:** Hardens curator LLM fork to be rollback-safe: removes `terminal` toolset so all skills-tree mutations are ledgered via `skill_manage` (#96962).

**Correctness:**
- `agent/curator.py:1950` — `enabled_toolsets=["skills"]` (was `["skills", "terminal"]`). Prompt updated to instruct via `skill_manage write_file/remove_file/delete` instead of shell `mv/cp/rm`. Eliminates hollow-archive bug where `mv` stripped package before snapshot.
- `_run_llm_review` dry_run path and consolidation instructions updated consistently. No terminal fallback remains.
- Tests updated to assert `skills` only.

**Non-blocking:**
- Prompt now says “NO terminal access” — verify no other code path still calls terminal for skills; grep confirms.

**Verdict:** LGTM.

kshitijk4poor:
Closing as **implemented on main**.

This landed via #99491 (commit `37ae0e1d39`): `terminal` is removed from the consolidation fork's `enabled_toolsets` (now `["skills"]`), the prompt steers re-homing through the ledgered `skill_manage` surface, and the three pinning tests are on main. Issue #96962 is closed. #99491 credits you for the minimal implementation that shipped.

Thanks @zengzheqing — this was the right-sized fix for the cluster.

PR #98698, opened 2026-08-30
fix(curator): restore complete skill packages on ledger rollback (#96962)

zengzheqing:
## Bug Description

Issue #96962: `hermes curator rollback` restored a hollow skill — `SKILL.md` back, `references/` gone. Root cause (verified in the issue thread): consolidation re-homed the skill's support files into an umbrella with un-ledgered terminal moves, so the archive/delete entry that followed captured only what was left on disk (`files: 1`). The support files were recoverable only by hand out of the pre-run `.curator_backups` tar.

## Fix

The package-destroying captures (`delete` / `archive` / `purge`) complete themselves from the newest curator `skills.tar.gz`:

- `fill_snapshot_from_curator_backup()` unions in package files missing from the disk capture. **Disk hashes win** — the backup fills only absent paths. Tar members escaping the package prefix (absolute paths, `..` traversal) are rejected; every fill target must stay under `skills/` and HERMES_HOME.
- Wired at the four capture sites: `skill_manage` delete, `archive_skill`, `purge`, and `record_mutation` (which auto-completes delete/archive/purge regardless of caller).
- `rollback_entry` applies the same fill, so **hollow entries recorded before this fix still restore the complete package**.
- Fill targets address where rollback restores: under the captured root (purge fills into `.archive/<name>/`, not the live tree).
- Fail-open by design: fill failures are swallowed and the original capture stands — this is completeness, not a gate.

Backup-id format and arcname layout were verified against `agent/curator_backup.snapshot_skills` (arcnames are relative to `skills/`; the id regex matches `_ID_RE` exactly). The pre-run snapshot is taken before the LLM pass, so the fill sees support files at their pre-re-home locations.

## Test Plan

Five tests in `tests/tools/test_skill_ledger.py`:

- [x] Incident shape end-to-end: create skill with `references/extra.md` → snapshot tarball → re-home out of tree → delete → entry captures BOTH files → rollback restores both
- [x] Historical hollow entry (hand-recorded `files: 1`) + tarball → rollback restores the full package
- [x] No backup present → degrades to old behavior, delete→rollback round trip intact
- [x] Disk hash wins over stale backup copy
- [x] Fill is live AND traversal members rejected (positive + negative assertion in one test — a silently inert fill cannot pass)

Discriminator verified: short-circuiting the fill function to identity turns all four fill tests red; restore → 19/19 green. Full regression: 123/123 across `test_skill_ledger.py`, `test_skill_manager_tool.py`, `test_skill_usage.py`, `test_curator_backup.py`, `test_curator_archive_prune.py`, `test_curator_run.py` via `scripts/run_tests.sh`, first try.

## Risk Assessment

Low-moderate. Fill only adds missing paths to package-destroying captures; all other actions capture disk state exactly as before. Tar read failures degrade silently to today's behavior.

## Related

- Fixes the recovery half of #96962; companion to the prevention-layer PR in this series (#98697).
- Mechanism credit: this is the backup-fill recovery approach assessed in the public comment on #96962 — same mechanism as #96980, reimplemented minimal with discriminating tests; if maintainers prefer #96980's version, this PR converts to review support.

### DIFF
--- a/hermes_cli/curator.py
+++ b/hermes_cli/curator.py
@@ -653,7 +653,9 @@ def _cmd_purge(args) -> int:
 
     purged = 0
     for p in sorted(candidates):
-        before = skill_ledger.capture_before(p)
+        before = skill_ledger.capture_before(
+            p, complete_package=True, skill=p.name
+        )
         try:
             shutil.rmtree(p)
         except OSError as e:
--- a/tests/tools/test_skill_ledger.py
+++ b/tests/tools/test_skill_ledger.py
@@ -365,3 +365,196 @@ def test_user_actor_override(ledger_env):
         skill_ledger.reset_ledger_actor(tok)
     entry = skill_ledger.get_entry(entry_id)
     assert entry["actor"] == "user"
+
+
+# ---------------------------------------------------------------------------
+# Package-completeness fill from the newest curator backup (issue #96962)
+# ---------------------------------------------------------------------------
+
+
+def _write_skills_tarball(home: Path, files: dict, stamp: str = "2026-08-01T00-00-00Z"):
+    """Write a curator-shaped ``skills.tar.gz`` under *home* (arcnames are
+    relative to skills/, exactly like agent.curator_backup.snapshot_skills)."""
+    import io
+    import tarfile
+
+    snap = home / "skills" / ".curator_backups" / stamp
+    snap.mkdir(parents=True, exist_ok=True)
+    tar_path = snap / "skills.tar.gz"
+    with tarfile.open(tar_path, "w:gz") as tf:
+        for rel, content in files.items():
+            data = content.encode("utf-8") if isinstance(content, str) else content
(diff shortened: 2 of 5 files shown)


alt-glitch:

Duplicate of #96980: both implement backup-assisted complete skill-package capture and restoration for the #96962 hollow-rollback failure.

teknium1:
Your two PRs (#98697 + #98698) have been composed into salvage PR #99491, cherry-picked onto current main with your commit authorship preserved in git log. #98698 is the later of the two duplicate recovery implementations (after @HexLab98's #96980, submitted first) — your version was picked for the purge-destination correctness, rollback-time re-validation, and traversal-rejection tests, and both of you are credited in the salvage PR body. Closing this one as superseded by the salvage vehicle; nothing further is needed from you. Thanks for the rigorous root-cause correction on #96962 and the discriminator-verified tests — both carried over intact.

PR #86628, opened 2026-08-15
feat(dashboard): expose skill deletion on the Skills page (port of centaur#1393)

teknium1:
## Summary
The dashboard Skills page can now delete agent/hand-authored skills — closing the last gap in its SKILL.md lifecycle (create and edit already existed; removal previously required SSH + `rm -rf`).

Ported from [paradigmxyz/centaur#1393](https://github.com/paradigmxyz/centaur/pull/1393) ("feat: expose skill deletion"), which wired skill archival through their console/CLI surface. Adapted to our architecture: instead of a new endpoint stack, `DELETE /api/skills` reuses the agent's guarded `_delete_skill` path, so every existing safety net applies to dashboard deletes too.

## Ours vs theirs
| | Centaur | This port |
|---|---|---|
| Surface | `centaur-skills` CLI + sandbox DELETE endpoint | Dashboard Skills page + `DELETE /api/skills` |
| Semantics | Archival (recoverable) | Hard delete for user-directed action (matches `skill_manage` foreground semantics); curator-pass deletes stay archival |
| Guards | Owner-only authz | Pinned-skill guard (→ 409), org-mirror guard, skills-root/symlink defense-in-depth, profile scoping |
| Scope | All skills, owner-gated | Only `provenance == "agent"` skills get the affordance — bundled skills are re-seeded by updates, hub skills uninstall via the Hub tab |

## Changes
- `hermes_cli/web_routers/skills.py`: `DELETE /api/skills?name=&profile=` → `_delete_skill(name, absorbed_into="")`; 404 not-found, 409 pinned, 400 guard refusals; clears the skills prompt cache on success.
- `web/src/pages/SkillsPage.tsx`: hover trash button on agent-authored rows, `DeleteConfirmDialog` confirm, optimistic list update.
- `web/src/lib/api.ts`: `deleteSkill()`; `SkillInfo.provenance` typed (the backend already returned it).
- `web/src/i18n/en.ts` + `types.ts`: optional `skills.deleteSkill*` keys (all other locales fall back via `defineLocale`).
- `tests/hermes_cli/test_web_server_skill_editor.py`: delete happy path, unknown-skill 404, profile scoping, pinned 409, 401 without session token.

## Validation
| Check | Result |
|---|---|
| `tests/hermes_cli/test_web_server_skill_editor.py` | 17/17 pass (13 existing + 4 new) |
| `tsc -p web --noEmit` | clean |
[...]

PR #58288, opened 2026-07-04
Add dashboard skill deletion

ap1969:
## Summary
- add authenticated DELETE /api/skills/{name} endpoint wired through the existing skill manager delete guard
- remove deleted skills from disabled-skill config and invalidate the prompt cache
- add dashboard API client support plus Skills page delete button and confirmation dialog
- cover delete success, profile scoping, missing skills, pinned-skill refusal, and auth protection

## Verification
- `python -m pytest tests/hermes_cli/test_web_server_skill_editor.py -q` → 22 passed
- `python -m pytest tests/hermes_cli/test_web_server_skill_editor.py tests/tools/test_skill_manager_tool.py -q -o 'addopts='` → 129 passed
- `python -m ruff check hermes_cli/web_server.py tests/hermes_cli/test_web_server_skill_editor.py` → passed
- `npm ci` → installed dependencies, 0 vulnerabilities
- `npm run typecheck` → passed
- `npm run build` → passed

PR #98700, opened 2026-08-30
Closed

algodondev:
Closed.


alt-glitch:

Related to #58288 and #86628, which already implement the same dashboard skill-deletion capability with UI and different endpoint contracts. Please consolidate rather than merge overlapping APIs.

PR #65603, opened 2026-07-16
fix(insights): include auxiliary usage in overview token totals

teknium1:
## Summary

`hermes insights` overview token totals now include auxiliary model usage, fixing the top-line undercount reported in #58592 and requested in #9979.

Root cause: `_compute_overview` summed token counters from the `sessions` rows (main-loop usage only), while the per-model breakdown below it already included auxiliary usage rows (task dimension added in #65537) plus reconciled residuals — so the overview disagreed with its own per-model table and with provider bills whenever compression/vision/title calls burned tokens.

## Changes

- `agent/insights.py`: when the per-model breakdown is available, derive the overview's input/output/cache token totals from it — the exact pattern `total_cost` already used one line above. 12 lines.
- `tests/hermes_state/test_aux_usage_accounting.py`: 2 new tests — aux tokens included in overview totals; no double-count across incremental CLI deltas + gateway absolute overwrites + aux rows.

## Validation

| | Before | After |
|---|---|---|
| Overview total_input (1000 main + 5000 compression + 300 vision) | 1000 | 6300 |
| Overview vs per-model table | disagreed | consistent |
| Gateway absolute + aux mix | — | 2040 (no double count, verified) |

- 77 tests green across `test_aux_usage_accounting.py` + `test_insights.py`.
- E2E with real SessionDB: seeded CLI-style (incremental) and gateway-style (absolute overwrite) sessions plus aux rows; totals exact with zero double-counting.

Closes #58592. Closes #9979.
Related: #60758 (@Bartok9 identified the compression-summarizer undercount and proposed folding summarizer usage into session accounting — the recording half landed via #65537's chokepoint, this PR completes the insights half).

## Infographic

Insights aux totals

PR #98701, opened 2026-08-30
fix(insights): include aux usage in platform breakdown and Most tokens

Gabrielnkl:
## Summary

`_compute_overview()` was fixed in #65603 to include `session_model_usage` rows (including auxiliary usage such as compression, vision, and titles), but `_compute_platform_breakdown()` and `_compute_top_sessions()` still relied only on the session aggregate counters.

This caused Insights views to disagree about token totals and could cause **Most tokens** to select the wrong session.

### Changes

* **Platform breakdown**

  * Passes `cutoff` and `source` from `generate()` into `_compute_platform_breakdown()`.
  * Aggregates `session_model_usage` by session/source so auxiliary usage is included.
  * Reconciles per-token residuals against the session aggregate counters, matching the existing model-breakdown/overview accounting pattern.
  * Falls back to session counters when called without a window or when no model-usage rows are available.
  * Fixes `source` parameter shadowing.

* **Most tokens**

  * Passes `cutoff` and `source` into `_compute_top_sessions()`.
  * Ranks sessions using input + output + cache read + cache write + auxiliary usage.
  * Uses the same per-token residual reconciliation as the platform/model breakdowns.
  * Falls back to session counters when no model-usage rows are available.

### Why the per-token residual matters

Residual reconciliation must be performed independently for each token field rather than on the combined total.

For example, if session counters contain:

* input: 1000
* output: 500

while `session_model_usage` contains:

* input: 1200
* output: 400

the correct reconciled total is:

`1200 + 500 = 1700`

A single `max(session_total, usage_total)` would incorrectly produce `1600`.

### Verification

Verified that:

* `overview.total_tokens == platform token sum`
* **Most tokens** selects the correct session when cache/auxiliary usage changes the ranking.
* The per-token residual case above produces `1700`.
* The original auxiliary + cache reproduction is reconciled correctly.
* `tests/agent/test_insights.py` passes.
* `tests/hermes_state/test_aux_usage_accounting.py` passes.

Final implementation: **178 additions / 31 deletions** in `agent/insights.py`.

### DIFF
--- a/agent/insights.py
+++ b/agent/insights.py
@@ -763,6 +763,29 @@ class InsightsEngine:
         result.sort(key=lambda x: (x["total_tokens"], x["sessions"]), reverse=True)
         return result
 
+    def _add_session_token_totals_to_platform_data(
+        self,
+        platform_data: Dict[str, Dict[str, Any]],
+        sessions: List[Dict],
+    ) -> None:
+        """Add session-level token aggregates to platform_data.
+
+        Sums input_tokens, output_tokens, cache_read_tokens, cache_write_tokens
+        and sets total_tokens = sum of those four fields.
+        """
+        for s in sessions:
+            src = s.get("source") or "unknown"
+            d = platform_data[src]
+            inp = s.get("input_tokens") or 0
+            out = s.get("output_tokens") or 0
+            cr = s.get("cache_read_tokens") or 0
+            cw = s.get("cache_write_tokens") or 0
+            d["input_tokens"] += inp
+            d["output_tokens"] += out
+            d["cache_read_tokens"] += cr
+            d["cache_write_tokens"] += cw
+            d["total_tokens"] += inp + out + cr + cw
+
     def _compute_platform_breakdown(
         self,
         sessions: List[Dict],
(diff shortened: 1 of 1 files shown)


Enough1122:

**Overall:** Fixes platform-breakdown token stats to include auxiliary rows (vision/compression) via `session_model_usage`, aligning with overview (#65603) and avoiding variable shadowing bug.

**Correctness:**
- `agent/insights.py:191` — `_compute_platform_breakdown` now takes `(sessions, cutoff, source)` and aggregates from `usage_rows` when available; falls back to session counters when `cutoff is None` or no rows. Correctly computes residuals for unknown platforms. Renames inner `src` to avoid shadowing param `source`.
- `_compute_top_sessions` similarly threaded.

**Non-blocking:**
- Double-loop fallback repeats same token summation; consider helper to deduplicate. `total_tokens` now `inp+out+cr+cw` consistent.

**Verdict:** LGTM.

Gabrielnkl:
Addressed the code duplication by extracting the shared session token aggregation logic into `_add_session_token_totals_to_platform_data()` and reusing it in both fallback paths.

Tests pass locally.

PR #98702, opened 2026-08-30
fix(desktop): persist live clarification cards

tranhieu204-web:
## Summary

Fix Hermes Desktop clarification cards disappearing while a session still shows **Needs your input**, especially after long-session hydration, runtime/stored-session rebinding, switching sessions, reconnect replay, and settlement from another renderer.

The repair makes the pending backend request authoritative and keeps its transcript projection, runtime identity, attention state, and settlement lifecycle coherent and idempotent.

### Key changes

- Reconcile exactly one unresolved clarification row/card from the authoritative pending request.
- Move pending clarification state exactly once when a stored session receives a new runtime ID.
- Keep distinct request-ID epochs isolated even when their question text is identical.
- Bind the model tool-call ID to the gateway request ID as a renderer-only, one-shot, same-epoch alias, allowing another renderer's answer to settle locally without resurrecting stale input.
- Preserve failed answer/Skip recovery and existing single/batch payload behavior.
- Add a production-shaped 741+ row Electron regression covering cold resume, newest-tail hydration, session switching, one answer, attention clear, and turn continuation.

## Test plan

- [x] Focused clarification/session/store tests: **202/202**
- [x] Desktop UI project: **5156/5156**
- [x] TypeScript typecheck
- [x] Changed-path lint: **0 errors**
- [x] Production Desktop build
- [x] Long-session clarification Electron E2E
- [x] Batch clarification Electron E2E
- [x] Independent different-family source review
- [x] Packaged Windows Desktop deployment and held-out live acceptance: switch away/back, exactly one answerable card, answer once, turn continues

## Notes

A separate pre-existing transcript backfill issue was discovered while developing the long-session regression: after two older pages, **Show earlier messages** may remain offered without delivering more history. That path is not changed by this PR and is intentionally deferred.


alt-glitch:

Related to #98451: both protect pending clarification cards during Desktop activation/hydration, while this PR covers a broader lifecycle. Please consolidate overlapping state ownership before merge.

PR #98716, opened 2026-08-30
fix(oauth): surface actionable guidance when sign-in windows lapse

### LINKED ISSUE #98682: Pain cluster: Nous Portal OAuth flows dead-end after sign-in (PKCE callbackUrl dropped ~Aug 27 + device-flow approval gap) with zero client-side error
## Cluster (weekly user-pain miner)

**Class:** BUG (portal-side regression) + UX GAP (no client-side error surface)
**Occurrences this week:** 2 direct Discord threads + 1 adjacent; plus prior tracked device-flow gap #47950.

Since ~Aug 27, "Sign in with Nous" against self-hosted gateways (Desktop + native PKCE clients) never completes: Portal sign-in succeeds but the post-sign-in handler lands on `portal.nousresearch.com/?signin=1&callbackUrl=%2Foauth%2Fauthorize...` and never resumes `/oauth/authorize`, so the gateway `/auth/callback` is never hit. Gateway log shows only `native_authorize_start`. Worked Aug 23–27 for the same user, so this is a Portal deploy regression, distinct from (but rhyming with) the device-flow approval gap in #47950 where `/manage-subscription?user_code=...` never presents the approval step.

### Instances
1. **Discord — "Portal OAuth for self-hosted dashboard stops after sign-in, never redirects"** (Aug 29, detailed repro, v0.20.5 `791e2ae`): https://discord.com/channels/1053877538025386074/1543231358791589979
   > "Browser lands on `https://portal.nousresearch.com/?signin=1&callbackUrl=%2Foauth%2Fauthorize...`. Stays there. No approval page, no redirect to redirect_uri, no error."
2. **Discord — "Can't authenticate Nous Portal - just hangs"** (Aug 18, device-code flow, Mac Desktop + CLI, Chrome and Safari): https://discord.com/channels/1053877538025386074/1539232123876151396 — same shape as #47950 (`/manage-subscription` shows the plan page, never the approval prompt).
3. **Adjacent — Discord "Cron scheduler stopped firing — agent-cron provisioning rejected (403 invalid_client)"** (Aug 16): https://discord.com/channels/1053877538025386074/1538604777225855018 — portal OAuth client provisioning failing for a different grant.

### Why file here
The Portal fix itself is server-side, but hermes-agent owns two actionable halves:
- **UX GAP:** both flows fail with *zero* user-visible error — Desktop sits on "waiting for you to authorize" until timeout; the dashboard PKCE flow shows nothing at all. The poller knows the code expired / callback never came; it should say so and suggest the fallback (API key / retry / status link).
- **Tracking:** #47950 covers only the device-code approval page. The dashboard code+PKCE `callbackUrl` drop is a separate Portal route and a fresh regression (~Aug 27) — the Portal team needs a distinct handle.

### Proposed fix
1. Portal: post-sign-in handler must resume the `callbackUrl` (`/oauth/authorize?...`) instead of dropping to home; device-flow half stays tracked in #47950.
2. hermes-agent: when device-code polling expires or the native PKCE wait times out, surface an explicit error with the elapsed state ("Portal never redirected back — this is usually a Portal-side sign-in issue; see status / try `hermes auth login --api-key`").

Related: #47950, #81318 (hosted dashboard invalid_grant), #61270 (gateway startup blocks on Portal OAuth exhaustion with no visible error — same silent-failure UX class).

*Filed by the weekly user-pain miner cron.*

kokhlo:
## Summary

**Root cause.** Both OAuth UIs race their own local `expires_in` countdown against the backend poller, and the countdown wins the wrong way:

- **Web (`OAuthLoginModal.tsx`)** — when the countdown hit zero, the tick immediately flipped the modal to a bare `t.oauth.sessionExpired` ("Session expired. Click Retry…"). The backend poller already holds the actionable message for exactly this case — `_nous_device_auth_timeout_message` in `hermes_cli/auth.py` ("Timed out waiting for device authorization… Portal sign-in is required before the device code can be approved… finish signing in at the Portal, then retry"), surfaced through the session's `error_message` — but the client's 1s countdown fired first and replaced it. Result: a dead-end error with zero guidance, the exact "zero client-side error" UX gap in #98682.
- **Desktop (`store/onboarding.ts`)** — `pollSession()` had no local expiry at all: a session the user abandoned kept polling every 2s forever (bounded only by the gateway's 15-min session GC), with the UI stuck on "Waiting for you to authorize…".

**Fix.**
- Web: at local lapse, poll the session one last time. If the backend has an `error_message` (the enriched timeout text), show it; if the backend still says `pending` (local clock ran ahead), keep polling instead of killing a live session; otherwise fall back to new guidance naming the common cause (sign-in page stalled in the opened tab — server-side; finish sign-in there, then Retry; API-key/CLI fallback). The countdown updater is now pure (decrement only); a separate effect owns the lapse. Retry also resets the countdown so a fresh attempt can't insta-lapse.
- Desktop: `schedulePollExpiry()` lapses the flow when its own `expires_in` window passes, with the same actionable message; the timer is cleared through `clearPoll()` so cancel/success/any flow exit disarms it.
- Both messages localize: `sessionExpiredNoError` (web en + types; other locales fall back via `defineLocale`) and `onboarding.signInExpired` (desktop en/types + the full locales zh/ja/zh-hant/ar).

The Portal-side `callbackUrl` drop itself is server-side and stays with the Portal team — this covers the hermes-agent client half the issue files against this repo.

## Testing

- `web`: `npx vitest run src/components/OAuthLoginModal.test.tsx` — **4 passed** (new: backend message surfaced at lapse; keeps polling on backend-pending/clock-skew; PKCE fallback guidance; Retry starts fresh). `npm run typecheck` clean, `npx eslint` on touched files 0 errors (1 pre-existing-class `set-state-in-effect` warning, same as `App.tsx` on main).
- `apps/desktop`: `npx vitest run src/store/onboarding.test.ts` — **17 passed** (15 existing + 2 new: expiry lapses with actionable guidance while still pending; window-open keeps polling and cancel disarms the expiry). `npm run typecheck` clean, `npx eslint` on touched files 0 problems.

Fixes #98682 (client-side half)

### DIFF
--- a/apps/desktop/src/i18n/ar.ts
+++ b/apps/desktop/src/i18n/ar.ts
@@ -2163,6 +2163,8 @@ export const ar = defineLocale({
     connectedProvider: provider => `تم ربط ${provider}`,
     connectedPicking: provider => `تم ربط ${provider}. جار اختيار نموذج افتراضي...`,
     signInFailed: 'فشل تسجيل الدخول. حاول مرة أخرى.',
+    signInExpired:
+      'انتهت مهلة انتظار التفويض. السبب الأكثر شيوعًا هو تعطّل صفحة تسجيل الدخول في تبويب المتصفح (مشكلة من جهة الخادم) — أكمل تسجيل الدخول هناك ثم أعد المحاولة. إذا استمر الفشل، استخدم مفتاح API أو واجهة سطر الأوامر بدلاً من ذلك.',
     pickDifferentProvider: 'اختر مزوداً آخر',
     signInWith: provider => `تسجيل الدخول عبر ${provider}`,
     openedBrowser: provider => `فتحنا ${provider} في المتصفح.`,
--- a/apps/desktop/src/i18n/en.ts
+++ b/apps/desktop/src/i18n/en.ts
@@ -2780,6 +2780,8 @@ export const en: Translations = {
     connectedProvider: provider => `${provider} connected`,
     connectedPicking: provider => `${provider} connected. Picking a default model...`,
     signInFailed: 'Sign-in failed. Try again.',
+    signInExpired:
+      'Sign-in expired waiting for authorization. This usually means the sign-in page stalled in the opened tab (server-side issue) — finish signing in there, then try again. If it keeps failing, use an API key or the CLI fallback instead.',
     pickDifferentProvider: 'Pick a different provider',
     signInWith: provider => `Sign in with ${provider}`,
     openedBrowser: provider => `We opened ${provider} in your browser.`,
--- a/apps/desktop/src/i18n/ja.ts
+++ b/apps/desktop/src/i18n/ja.ts
@@ -2422,6 +2422,8 @@ export const ja = defineLocale({
     connectedProvider: provider => `${provider} が接続されました`,
     connectedPicking: provider => `${provider} が接続されました。デフォルトモデルを選択中...`,
     signInFailed: 'サインインに失敗しました。再試行してください。',
+    signInExpired:
+      '承認待ちでタイムアウトしました。多くの場合、開いたタブのサインインページが止まっている（サーバー側の問題）ためです。そのページでサインインを完了してから再試行してください。解決しない場合は API キーまたは CLI を利用してください。',
     pickDifferentProvider: '別のプロバイダーを選択',
     signInWith: provider => `${provider} でサインイン`,
     openedBrowser: provider => `${provider} をブラウザーで開きました。`,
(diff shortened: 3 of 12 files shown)


teknium1:
Thanks — landed on main via #100434 with authorship preserved: lapsed OAuth sign-in windows now surface actionable guidance instead of hanging silently (the repo-side UX half of #98682; the portal-side callbackUrl fix remains with the Portal team). Closing as superseded by the merged salvage.

PR #23963, opened 2026-05-11
feat(agent): trigger-based skill auto-loading from SKILL.md frontmatter

it-helloprint:
## Problem

Skills are only loaded when the agent *recognises* the match from a one-line description. If a skill description does not surface for a given trigger condition, the knowledge is never applied — even if the exact fix is documented.

Real example: a `PHP ParseError: syntax error, unexpected token "if"` in CI was debugged from scratch even though `laravel-pitfalls` had the exact fix documented. The one-line description did not match the error pattern.

## Solution

Add a `triggers:` key to SKILL.md frontmatter. When regex patterns match the incoming user message, the skill is automatically injected into context **before the agent reasons** — no recognition gap.

```yaml
triggers:
  - 'PHP Parse error.*syntax error'
  - 'mergeable.*CONFLICTING'
  - 'inStore\(null\)'
```

Patterns are matched case-insensitively. Invalid regex is skipped without error. At most 5 skills are auto-loaded per turn (configurable via `MAX_AUTO_LOADED_SKILLS`). Results are mtime-cached to avoid repeated disk reads in long gateway sessions.

### `agent/skill_trigger_loader.py` (new, ~170 LOC)
- `get_triggered_skills(text)` — scan all SKILL.md files, match triggers, return full contents
- `format_triggered_skills_block(skills)` — format for context injection
- mtime-based cache, profile-aware skills home, fail-safe throughout

### `run_agent.py` (+25 LOC)
- Hook at start of `run_conversation()`, first turn only
- Injects matched skills as a `system` message before the user turn
- Wrapped in `try/except` — any failure is logged at DEBUG, never raised

### `tests/agent/test_skill_trigger_loader.py` (new, 14 tests)
- Blank input, case-insensitive match, no-match, skip-no-triggers
- Multiple matches, per-skill cap, single-load-per-skill
- Invalid regex graceful skip, nonexistent skills home
- mtime cache correctness, format functions

All 14 tests pass.

## Fail-safe design

The hook in `run_agent.py` is wrapped in a broad `except Exception` — a bug in the trigger loader can never crash or slow the agent loop. Failures are logged at `DEBUG` level only.

## Backwards compatibility

Fully additive. Skills without a `triggers:` key behave exactly as before. No config changes required.

PR #98718, opened 2026-08-30
feat(skills): add onload hook and HERMES_HOME template var for environment-triggered skill loading

CaptainGimpy:
## Summary

Two additive changes to the skill system that extend it without breaking existing behaviour.

### 1. Onload hook

Skills can now declare an `onload:` frontmatter field pointing to a script ( or ). At session init the script runs with the resolved model/provider in env; if it prints `INJECT` the skill body is pre-loaded into the system prompt volatile tier as `## Onload: <name>` — before the model sees any user message.

This lets **environment-triggered skills** fire without depending on conversational keyword matching:
- Model-specific behavior mandates (e.g. a DeepSeek mandate that constrains tool-use patterns)
- Safety rules that must load early
- Platform-specific protocols
- Compliance guardrails

### 2. `HERMES_HOME` template variable

`${HERMES_HOME}` now resolves in SKILL.md content alongside the existing `${HERMES_SKILL_DIR}` and `${HERMES_SESSION_ID}` vars. This lets cross-skill references use an absolute anchor (e.g. `${HERMES_HOME}/skref/<skill>/file`) instead of fragile relative paths.

## Changes

| File | Δ | Description |
|------|---|---|
| `agent/skill_preprocessing.py` | +80/-2 | Added `run_onload_script()`, `HERMES_HOME` template var support |
| `agent/system_prompt.py` | +57/-0 | Onload runner: iterates entries, runs scripts, injects bodies on `INJECT` |
| `agent/prompt_builder.py` | +38/-8 | Extracts `onload` + `rel_path` from frontmatter, v3 snapshot bump, returns `(index, onload_entries)` tuple |

## Design

Both changes are purely **additive** — no existing skill breaks, no existing reference breaks. The session init path gains only a new injector stage in the volatile tier builder:

## Testing

- `onload` hook verified: check_model.py runs at session init, detects DeepSeek provider, returns `INJECT`, mandate body lands in system prompt
- `HERMES_HOME` var verified: `${HERMES_HOME}/skref/<skill>/file` resolves and loads correctly
- Backward compat: skills without `onload:` field load identically, old v2 snapshot cache entries are auto-wrapped to the new tuple contract

Co-authored-by: CaptainGimpy <captaingimpy@users.noreply.github.com>

### DIFF
--- a/agent/prompt_builder.py
+++ b/agent/prompt_builder.py
@@ -1593,7 +1593,7 @@ _SKILLS_PROMPT_CACHE: OrderedDict[tuple, str] = OrderedDict()
 _SKILLS_PROMPT_CACHE_LOCK = threading.Lock()
 # v2: entries gained org provenance fields (org_id/org_author/rel_dir) for M2
 # org-shared skills; older snapshots are discarded and rebuilt.
-_SKILLS_SNAPSHOT_VERSION = 2
+_SKILLS_SNAPSHOT_VERSION = 3
 
 
 def _skills_prompt_snapshot_path() -> Path:
@@ -1729,6 +1729,8 @@ def _build_snapshot_entry(
         "description": description,
         "platforms": [str(p).strip() for p in platforms if str(p).strip()],
         "conditions": extract_skill_conditions(frontmatter),
+        "onload": str(frontmatter.get("onload", "") or ""),
+        "rel_path": str(rel_path),
     }
     if org_id:
         entry["org_id"] = org_id
@@ -1830,7 +1832,7 @@ def build_skills_system_prompt(
     available_toolsets: "set[str] | None" = None,
     compact_categories: "frozenset[str] | None" = None,
     skills_dir_override: "Path | None" = None,
-) -> str:
+) -> tuple[str, list[dict]]:
     """Build a compact skill index for the system prompt.
 
     Two-layer cache:
@@ -1896,7 +1898,7 @@ def _build_skills_system_prompt_inner(
     available_toolsets: "set[str] | None",
     compact_categories: "frozenset[str] | None",
     project_dirs: "list[Path] | None" = None,
-) -> str:
+) -> tuple[str, list[dict]]:
     # Include the resolved platform so per-platform disabled-skill lists
     # produce distinct cache entries (gateway serves multiple platforms).
     _platform_hint = _current_session_platform_hint()
@@ -1916,6 +1918,11 @@ def _build_skills_system_prompt_inner(
         cached = _SKILLS_PROMPT_CACHE.get(cache_key)
         if cached is not None:
             _SKILLS_PROMPT_CACHE.move_to_end(cache_key)
+            # Old-format cached strings (v2 snapshots) get wrapped to
+            # maintain the new (index, onload_entries) contract.
+            if isinstance(cached, str):
+                cached = (cached, [])
+                _SKILLS_PROMPT_CACHE[cache_key] = cached
             return cached
 
     # ── Layer 2: disk snapshot ────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


alt-glitch:

Related to #23963: both add frontmatter-driven skill auto-loading, but this PR uses an environment-triggered onload script at session initialization rather than its trigger-matching mechanism. Please consider consolidating the intended model.

CaptainGimpy:
Re: your consolidation question — PR #98768 (/pull/98768) addresses issue #23963 with a redesigned trigger-based skill auto-loading mechanism that is orthogonal to the `onload:` hook in this PR.

**The unified model:**

| Key | Fires | Evaluated by | Injects via | Cache impact |
|-----|-------|-------------|-------------|--------------|
| `onload:` (this PR) | Session init | External .py/.sh script | Volatile system prompt | Zero (built-once) |
| `triggers:` (#98768) | Every turn | Regex on user message | api_content sidecar on user msg | Zero (sidecar is cache-safe) |

Two different problems, two different injection paths — no conflict, no consolidation needed. Both are additive frontmatter keys in SKILL.md that share the same `_build_snapshot_entry` snapshot pipeline (and both add `rel_path` there, so whichever merges second just drops the duplicate during rebase).

#98768 consumes snapshot entries from prompt_builder rather than running its own filesystem scanner, injects via the existing api_content sidecar (same channel as memory prefetch — cache-safe), and evaluates every turn with no first-turn-only guard.

andrexibiza (commented):
Reviewed exact head `8b626919af81966578a8a1a1bec2d00da8eab701` against live `main@4f22543509d1b91dc45bcb369447126c5eb14fb7`, the full three-file diff, current skill preprocessing/config defaults, existing prompt-builder tests/callers, exact-head workflows, and the adjacent trigger work in #98768 / lineage from #23963.

The underlying capability is useful, and `${HERMES_HOME}` substitution is a clean independent addition. The `onload` execution path is not safe to land in its current shape yet. I see four concrete blockers:

### P1 — `onload:` bypasses the existing executable-skill trust gate and inherits the entire Hermes process environment

Current `main` deliberately defaults `skills.inline_shell` to `false`, with the explicit rationale that skill-author content otherwise “runs on the host without approval” and should only be enabled for trusted skill sources. This head introduces a second skill-authored code-execution path, but `system_prompt.py` calls `run_onload_script()` for every visible skill carrying `onload:` with no config/provenance/trust check at all. `run_onload_script()` then starts from `os.environ.copy()` and adds only `HERMES_MODEL` / `HERMES_PROVIDER`, so the child receives every credential/secret/environment capability held by the Hermes process.

That changes a visible `SKILL.md` from data/instructions into ambient host execution authority. Visibility filtering (platform/env/disabled/conditions) is not an execution grant.

Required before landing: make executable onload behavior explicit opt-in / fail-closed at the central execution boundary, preserve source/provenance policy, and construct an allowlisted child environment rather than inheriting the full parent environment. Add a regression proving a visible skill with `onload:` is inert when execution is not authorized, plus a regression that parent-only secrets are absent from the child.

### P1 — `onload` script resolution can escape the skill root

`script_path = skill_dir / script_rel_path` accepts frontmatter-controlled absolute paths and `../` traversal. There is no `resolve()` + containment check before execution. A skill can therefore nominate an existing `.py`, `.sh`, or fallback-bash target outside its own directory; an in-tree symlink can escape the root for the same reason.

Required: reject absolute paths, resolve both root and candidate, require the resolved candidate to remain inside the resolved skill directory, and test `../`, absolute-path, and symlink-escape cases. This must fail before spawning any process.

### P1 — the public `build_skills_system_prompt()` contract is broken

This head changes `build_skills_system_prompt()` from `str` to `tuple[str, list[dict]]`, while current callers/tests outside `system_prompt.py` still consume it as a string. Existing tests such as `test_deduplicates_skills`, `test_compact_categories_demote_nested_and_miss_cache_separately`, `test_excludes_disabled_skills`, and `test_rebuilds_prompt_when_disabled_skills_change` call `.count(...)`, string membership, etc. on the public result. Those semantics are invalid on the new tuple.

#98768 is useful adjacent evidence here: it changes the inner snapshot builder to return `(result, visible_entries)` but deliberately keeps `build_skills_system_prompt()` returning the original string and exposes a separate `get_visible_skill_entries()` seam. The two PRs are **adjacent/complementary, not duplicates**, but they directly collide in `agent/prompt_builder.py` and currently propose different public contracts.

Required: preserve the public string contract and expose onload metadata through a separate bounded API/shared visible-entry owner. If #98768 lands first, rebase onto that seam; if this lands first, #98768 must recompose rather than reintroduce the same snapshot contract. Keep #23963 / it-helloprint credited as the original trigger-feature lineage; the `onload` feature remains CaptainGimpy’s distinct adjacent work.

### P1 — this exact object has no acceptance evidence and is based on a very old merge base

The reviewed commit’s parent / actual merge base is `74ad422d50e6d76e26ba5ce3b4d2a4d520e923dd`. Relative to current `main@4f22543509d1b91dc45bcb369447126c5eb14fb7`, the branch has one unique commit and is **1,258 main commits behind**. Main changed `agent/prompt_builder.py` by 823 lines and `agent/system_prompt.py` by 190 lines across that gap, so mergeability is not runtime proof.

Exact-head CI `33323374142`, Docker `33323373814`, and Nix `33323373809` all concluded `action_required` with **zero jobs**. The PR also adds no test file despite changing a public return type and introducing automatic subprocess execution.

Please rebase/recompose onto current main after the authority/path/API fixes, add focused regressions for the new execution boundary, run the existing prompt-builder/system-prompt/skill preprocessing suites, and obtain fresh exact-head CI/Docker/Nix for every submitted commit.

### Ownership / merge-order note

`agent/prompt_builder.py` is already beyond line 2,200, and both #98718 and #98768 are adding new skill-routing metadata/policy there. This is exactly the point where the snapshot metadata/consumer seam should become a bounded owner instead of growing another policy branch inside the godfile. That also gives the two features one collision-free place to compose without conflating their ownership.

I would keep the `${HERMES_HOME}` template substitution — it is small and orthogonal — and preserve the onload idea, but move code execution behind an explicit authority boundary and make path containment + environment minimization part of that contract. Once those are true and the recomposed exact head is green, this becomes a much stronger feature rather than a new ambient execution surface.

CaptainGimpy:
The original PR (#98718) was force-pushed during fix-up and got closed by GitHub — the reworked feature is now at **#98779**.

All four P1s from this review thread are addressed there:
- Trust gate: `skills.onload_enabled` config (default false), allowlisted child environment
- Path containment: absolute paths rejected, `resolve()` + `relative_to()` containment check
- Public API preserved: `build_skills_system_prompt()` returns `str`, `get_onload_entries()` separate seam
- Rebased onto current `main` with 104 passing tests

The consolidation question is also answered — #98768 (triggers) landed as an orthogonal frontmatter key using the same snapshot pipeline but with a different injection path (api_content sidecar, per-turn regex).

PR #88877, opened 2026-08-18
fix(desktop): stop workspace edits from reloading external preview tabs

NealZhouPanda:
Salvages #81503 (original author credit preserved via cherry-pick), rebased onto current main.

## Problem

Any completed file diff force-reloads **every** URL-kind preview tab in the right rail — browsing x.com loses scroll position and feed state whenever the agent edits any workspace file, even files completely unrelated to the open page (`use-preview-routing.ts` bumps one global reload counter; `preview-pane.tsx` reloads any `target.kind === 'url'` pane unconditionally).

## Fix

Gate workspace live-reload to origins that can plausibly depend on workspace files, via a new `isWorkspaceLiveReloadUrl()`:

- loopback (full 127.0.0.0/8, `::1`, `0.0.0.0`)
- RFC 1918 private ranges (10/8, 172.16/12, 192.168/16) and link-local (169.254/16) — LAN-served dev servers keep working
- `localhost`, `*.localhost`, `*.local`

External origins (x.com etc.) are never auto-reloaded by workspace edits. Manual refresh, the file-watcher path (`kind === 'file'`), and dev-server restart reloads are unaffected.

## Changes beyond the salvaged commit

Cold review flagged that a localhost-only allowlist would regress LAN-served dev previews, so the second commit widens the allowlist as above and adds boundary tests: userinfo smuggling (`http://user@evil.test`), just-outside ranges (`172.32.x`, `11.x`), invalid octets, case and trailing-dot normalization.

## Test plan

- Targeted vitest 49/49 — including a RED check that the external-tab test fails without the gate
- Full UI suite 4709/4709
- `tsc --noEmit`, eslint, prettier all clean

Closes #81487

PR #81503, opened 2026-08-08
fix(desktop): stop workspace edits from reloading external preview tabs

fangliquanflq:
## What does this PR do?

Workspace file edits no longer reload unrelated external pages in the Desktop preview pane. This preserves scroll position, feed state, and in-progress reading on general web tabs while retaining automatic reloads for local development previews.

### Symptom

Completing any workspace file edit increments the shared preview reload request. A URL preview such as `https://x.com` consumed that request unconditionally and reloaded even though the page was unrelated to the changed file.

### Impact

Desktop users browsing an external site in the preview pane can lose transient page state, including their scroll position and current feed state, whenever an agent edits a workspace file.

### Bug Cause

**Trigger:** `apps/desktop/src/app/chat/right-rail/preview-pane.tsx` workspace reload effect

**Causal chain:**
1. A completed file diff increments the shared workspace reload request.
2. Every URL-kind preview consumes the new request without checking the current webview origin.
3. The active external page reloads and loses transient browsing state.

**Why it is wrong:** The reload consumer classified only the preview target kind, conflating local development servers with general-purpose web tabs.

**Working sibling / contrast:** File previews use a file watcher scoped by watch id, and explicit preview-server restarts match the restarted server URL before reloading.

**Ruled out:** The explicit preview-server restart path is not responsible because its reload is gated by a match between the restart URL and the current URL.

### Fix

Classify workspace-live-reload URLs by their current webview hostname. Loopback hosts and `.local` development hosts continue to reload, while external, malformed, deceptive, and non-HTTP(S) URLs ignore workspace reload requests. The consumer checks the current URL so a local tab that navigates to an external page is also protected.

## Related Issue

Fixes #81487

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- `apps/desktop/src/lib/local-preview.ts` - add the local development URL classifier.
- `apps/desktop/src/lib/local-preview.test.ts` - cover loopback, `.local`, malformed, deceptive, and external URL cases.
- `apps/desktop/src/app/chat/right-rail/preview-pane.tsx` - gate workspace-triggered reloads on the current webview URL.
- `apps/desktop/src/app/chat/right-rail/preview-pane.test.tsx` - verify external tabs stay untouched while local previews still reload.

## How to Test

1. Run the focused Desktop preview tests:

```bash
npx vitest run src/lib/local-preview.test.ts src/app/chat/right-rail/preview-pane.test.tsx
```

2. Run Desktop static checks:

```bash
[...]

Issue #81487, opened 2026-08-08
[Bug]: Desktop: any workspace file edit force-reloads unrelated web tabs in the preview pane (x.com etc.)


NealZhouPanda:
## Summary

With any web URL open in the desktop right-rail preview pane (e.g. browsing x.com), **every agent file edit anywhere in the workspace force-reloads that page** — even when the edited file is completely unrelated (e.g. the agent editing a markdown note). On sites like X this loses scroll position, drops the feed, and interrupts whatever the user was reading.

## Expected behavior

Workspace-change live reload should only apply to tabs where it makes sense — localhost dev-server previews (the original intent), or at minimum tabs whose content could plausibly depend on workspace files. A general-purpose web tab should never be reloaded by an unrelated file edit.

## Actual behavior

Any completed file-diff event bumps a single global reload counter, and every open URL-kind preview tab reloads unconditionally.

## Root cause (current main)

1. `apps/desktop/src/app/session/hooks/use-preview-routing.ts` — on any completed file diff, if *any* open tab is URL-kind:
   ```ts
   if ($previewTabs.get().some(tab => tab.target.kind === 'url') && gatewayEventCompletedFileDiff(event)) {
     requestPreviewReload()
   }
   ```
   No localhost check, no "is this file related to that tab" check.

2. `apps/desktop/src/store/preview.ts` — `requestPreviewReload()` is one global counter for all tabs.

3. `apps/desktop/src/app/chat/right-rail/preview-pane.tsx` — the reload effect fires for **any** `target.kind === 'url'` pane when the counter changes:
   ```ts
   if (target.kind !== 'url') { return }
   ...
   reloadPreview()
   ```

## Suggested fix

- Restrict the auto-reload to `localhost` / `127.0.0.1` / `.local` origins (matching the dev-server live-reload intent), and/or
- Scope reloads per-tab to the file that actually changed (the file-watcher path at `preview-pane.tsx` ~L443 already does this correctly for `kind === 'file'`), and/or
- Add a config gate (e.g. `preview.live_reload: localhost|all|off`).

## Related

- #41735 (scroll position resets on file reload) — same reload path, different symptom.

teknium1:
Fixed on main by #120055 (8206d89b984 + 680d613a018): the in-app browser now live-reloads only pages served from loopback (localhost / 127.x / ::1 / 0.0.0.0), judged by the tab's live page, so an agent file edit no longer reloads an external site and wipes its state.

[...]

PR #98719, opened 2026-08-30
fix(desktop): rest every browser overlay at turn end

BearHuddleston:
## What does this PR do?

Stops the Desktop in-app Browser's element overlay from continuing to pulse after a turn becomes idle when the active Browser tab changed during that turn.

A page could keep flashing/scanning elements even though Hermes was idle. The trigger looked random because it depended on which Browser tab was active at the start and end of the turn.

### Root cause

`preview-mind.ts` sent both busy-edge stages through `activePreviewScriptRunner()`:

1. Tab A was active when `$busy` became true, so A received `think` and started its pulse interval.
2. The user selected tab B while the turn was running.
3. When `$busy` became false, only the now-active tab B received `rest`.
4. Tab A never received cleanup, so its pulse interval continued while the session was idle.

The fix keeps `think` scoped to the active page but broadcasts the idle `rest` cleanup to every mounted preview script runner.

## Related Issue

No matching issue or PR was found for the idle pulse/tab-switch lifecycle; the bug was reported directly.

The separate workspace-edit refresh symptom is already tracked by #81487 and has the authorship-preserving open fix #88877 (salvaged from #81503), so this PR intentionally does not duplicate it.

## Type of Change

## Changes Made

- `apps/desktop/src/app/chat/right-rail/preview-mind.ts` — keep active-page `think`, but rest all mounted preview overlays when the turn ends.
- `apps/desktop/src/app/chat/right-rail/preview-nudge.ts` — share nudge dispatch and add renderer-wide idle cleanup.
- `apps/desktop/src/app/chat/right-rail/preview-script-runner.ts` — expose the mounted runner set to idle cleanup.
- `apps/desktop/src/app/chat/right-rail/preview-mind.test.ts` — reproduce tab A thinking, tab B becoming active, and the busy→idle edge.

## How to Test

1. On current `origin/main`, the regression test fails because tab A's last overlay stage remains `"think"`:
   ```bash
   cd apps/desktop
   npx vitest run src/app/chat/right-rail/preview-mind.test.ts
   ```
2. Apply this patch and rerun the focused test; it passes.
3. Run the complete Desktop gate:
   ```bash
   npm run check
   ```

### Code

- [x] I've read the Contributing Guide and Desktop `AGENTS.md`
- [x] My commit message follows Conventional Commits
- [x] I searched open and closed issues/PRs to avoid duplicates
- [x] This PR contains only the idle overlay lifecycle fix
- [x] The focused regression test was observed red before implementation and now passes
- [x] `npm run check` passes
- [x] Tested on Ubuntu 26.04

- [x] Documentation update — N/A; no user-facing API or setting changed
- [x] `cli-config.yaml.example` update — N/A; no config changed
- [x] `CONTRIBUTING.md` / `AGENTS.md` update — N/A; no workflow changed
- [x] Cross-platform impact considered — renderer-only state and guest-page script dispatch
- [x] Tool descriptions/schemas update — N/A; no model tool contract changed

### DIFF
--- a/apps/desktop/src/app/chat/right-rail/preview-mind.test.ts
+++ b/apps/desktop/src/app/chat/right-rail/preview-mind.test.ts
@@ -1,0 +1,52 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+const target = (url: string) => ({ kind: 'url' as const, label: url, source: url, url })
+
+describe('preview mind', () => {
+  beforeEach(() => {
+    vi.resetModules()
+  })
+
+  afterEach(async () => {
+    const { setBusy } = await import('@/store/session')
+    const { closeRightRail } = await import('@/store/preview')
+
+    setBusy(false)
+    closeRightRail()
+  })
+
+  it('rests the overlay that started thinking when another browser tab becomes active', async () => {
+    const { $rightRailActiveTabId, selectRightRailTab } = await import('@/store/layout')
+    const { newBrowserTab, openPreview } = await import('@/store/preview')
+    const { setBusy } = await import('@/store/session')
+    const { registerPreviewScriptRunner } = await import('./preview-script-runner')
+
+    openPreview(target('https://example.com'))
+
+    const firstTabId = $rightRailActiveTabId.get()!
+    const firstRunner = vi.fn(async (_code: string) => 'ok')
+    const unregisterFirst = registerPreviewScriptRunner(firstTabId, firstRunner)
+
+    newBrowserTab()
+
+    const secondTabId = $rightRailActiveTabId.get()!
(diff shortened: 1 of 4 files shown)


Enough1122:

**Overall:** Fixes stranded “thinking” pulse when switching preview tabs; `rest` must hit all overlays.

**Correctness:**
- `preview-mind.ts:26` — now `if(busy) nudgeOverlay('think') else restOverlays()`; `restOverlays()` iterates `previewScriptRunners()` and `nudgeRunner(run,"rest")` for each. New `preview-nudge.ts:42` + test `preview-mind.test.ts` verifies first tab's runner gets `rest` after switch to second tab.
- `think` stays active-tab only; `rest` is broadcast cleanup.

**Non-blocking:**
- None.

**Verdict:** LGTM.

PR #98583, opened 2026-08-30
fix(gateway): reconcile unrecorded streaming previews against visible prefix (#98552)

shali10:
## Problem

Fixes #98552.

When streaming Telegram/Discord/Slack responses, if the finalize edit fails or is not explicitly recorded (e.g. rate limiting, unrecorded preview), `delivered_final_matches()` in `gateway/stream_consumer.py` returned `None`.

On non-split paths, `gateway/run.py` treated `None` as legacy trust, causing `_stream_confirmed_final_delivery()` or `_content_delivered=True` to suppress the normal final send:
```text
Suppressing normal final send for session ...: final delivery already confirmed (streamed=True previewed=False content_delivered=True).
```

As reported in #98552, users were left with a truncated preview (e.g. 333 chars out of 624) and a stuck streaming cursor (`▉`), while the gateway silently swallowed the complete response.

## Solution

1. **Reconcile visible on-screen prefix**: In `GatewayStreamConsumer.delivered_final_matches()`, when `_delivered_final_text` is `None`, inspect `self._visible_prefix().strip()`. If an on-screen prefix is visible and demonstrably differs from the target completed response, return `False` to reject delivery confirmation and allow the complete send to proceed.
2. **Explicit result boolean validation**: In `gateway/run.py` stale finalize reconciliation edit check, ensure `_reconcile_res` is not None and `getattr(_reconcile_res, 'success', False) is True` rather than defaulting truthy on non-None values.
3. **Regression tests**: Added unit and end-to-end tests in `tests/gateway/test_stale_finalize_suppression.py` verifying unrecorded previews with cursors correctly trigger full delivery.

## Verification

```bash
pytest tests/gateway/test_stale_finalize_suppression.py tests/gateway/test_suppression_contract_matrix.py tests/gateway/test_stream_final_contract.py
```
All 67 tests passed (with 1 pre-existing xfail).

Issue #82656, opened 2026-08-09
Gateway suppresses final send on interrupted stream — partial with ▉ cursor left as only delivery (streamed=True previewed=False content_delivered=True)


AtlasOmnia:
## Bug Description

On Telegram, an interrupted stream can leave only a partial message ending with the streaming cursor `▉` in the chat, while the gateway logs:

```
Suppressing normal final send for session <key>: final delivery already confirmed (streamed=True previewed=False content_delivered=True).
```

and the complete response is never delivered (or arrives ~20s later only via a separate path). On agent-to-agent channels this reads as message loss: the recipient only ever sees a truncated fragment ending with the cursor.

## Evidence

1. **Windows gateway**, after a session recovery/recreation event: the agent transcript contains the full ack text, but the chat only ever received a short partial ending with `▉`. Two further attempts froze after only a few characters.
2. **Windows gateway**: the suppression line above was logged for a ~200-char response; the first visible copy was a partial ending with `▉`; the full text arrived only ~20s later as a second message (not an edit).
3. **macOS gateway**: the same suppression for a ~2000-char response immediately after the max-iterations summary path.

## Steps to Reproduce

1. Run the Hermes gateway against a Telegram group (edit-based streaming, default cursor ` ▉`).
2. Trigger a stream interruption mid-generation: session recovery/recreation (ended-but-live routing key), max-iterations summary, or a model/API failure while the reply is being generated.
3. Observe: the first visible message is a partial ending with `▉`; the log shows `Suppressing normal final send ... (streamed=True previewed=False content_delivered=True)`; the complete text never arrives (or only much later via a second delivery).

## Expected Behavior

The complete final text is always delivered. A partial preview with a stuck cursor must never satisfy the "final delivery already confirmed" check — `previewed=False` with no verifiable delivered payload should never suppress the normal final send.

## Actual Behavior

Suppression fires with `previewed=False`; only a partial-with-cursor is visible.

## Suspected Cause

[...]

Issue #78541, opened 2026-08-04
Suppressing normal final send swallows complete replies on Telegram group/forum sessions (payload-less split-delivery flags final_content_delivered)


GTHell:
# Bug report: "Suppressing normal final send" swallows complete replies on Telegram group/forum sessions (stale `content_delivered` + `None` trust)

## Symptom
On **Telegram group / forum-topic sessions** (`session_key` contains `:group:`), the gateway
sometimes produces a complete response and then **never sends it**. The user sees nothing —
not even a partial message. The log shows the suppression firing *at the same instant* the
response is ready:

```
2026-08-04 16:22:20,939 INFO gateway.run: Suppressing normal final send for session agent:main:telegram:group:-1004492624436:1: final delivery already confirmed (streamed=True previewed=False content_delivered=True).
2026-08-04 16:22:20,940 INFO gateway.run: response ready: platform=telegram chat=-1004492624436 time=55.0s api_calls=2 response=1939 chars
```

The 1939-char response never reached the chat. This is not one-off: **165 suppressed events
with 100–3000 char responses over 48h across 3 different groups** on this host, all with the
same signature (`streamed=True previewed=False content_delivered=True`).

Same-host **DM sessions also suppress** (e.g. `...:dm:449277883`, 1383-char reply), but DM
delivery demonstrably works (users receive the reply — confirmed via follow-up messages
quoting them), so DMs have a delivery path that groups do not. The group path is the broken one.

## Repro (observed trigger sequence)
1. A turn produces a **short streamed flush** for the group session
   (`[Telegram] Flushing text batch agent:main:telegram:group:-1004492624436:1 (17 chars)`)
   or the user issues `/stop` / a model switch mid-turn.
2. The stream consumer marks `final_content_delivered=True` from that early flush
   (`stream_consumer.py` sets it when `chunks_delivered and tail_delivered`; the
   multi-message split-delivery branch then records **no payload** —
   `_turn_split_delivery = True; _delivered_final_text = None`).
3. The *next* complete response reaches `run.py`'s suppression check:
   `_content_delivered = bool(_sc and getattr(_sc, "final_content_delivered", False))`.
4. The staleness matcher `delivered_final_matches(_final)` is consulted, but for the
   split-delivery path there is **no recorded payload** → matcher returns `None` → code
   comment says *"None (no record / multi-message split delivery) keeps legacy trust"* →
   `_content_delivered` stays `True` → the real final answer is suppressed.

## Root cause
[...]

Issue #71643, opened 2026-07-25
[Bug]: Telegram streaming — successful finalize edit carries the stale preview text, `content_delivered=True` suppresses the complete send


derniesner:
### Bug Description

A streamed Telegram reply can arrive permanently truncated **while every Bot API call in the turn succeeds**. The final `editMessageText` carries the last streamed *preview* text (cursor stripped, MarkdownV2 applied) rather than the completed response. The gateway then sets `content_delivered=True` and suppresses the normal final send, so the missing tail is never delivered by any path.

This is distinct from the failed-finalize family already handled on `main`:

- #51828 / `04898631` recover a final delivery **after a failed edit** (flood control). That commit is present in this install and does not apply here — there is no failure to recover from.
- PR #52971 likewise targets "premature content-delivered flag on **failed** finalize edits".
- #69486 describes truncation via Codex stream supersession, with display streaming disabled — a different trigger.
- #49687 reports the same `final_content_delivered`-flagged-too-early mechanic on Feishu, where the remainder at least still went out through a rescue path. Here nothing follows.

The regression comment on #51828 (2026-07-22) describes this symptom from a different install and asks for coverage of "successful-looking streamed finalize but visibly partial Telegram message". This report adds the payload-level evidence for that request.

### Steps to Reproduce

1. Telegram **group** chat, `streaming.enabled: true`, `transport: auto`, `edit_interval: 0.8`.
2. `logging.level: DEBUG`, restart the gateway.
3. Ask something that produces a reply of a few hundred characters and keeps generating past the last preview edit (here: image + "describe this photo").
4. Compare the `text` parameter of the last `editMessageText` call against the `response ready: ... response=N chars` line for the same turn.

Reproduces reliably here; observed on three separate days.

### Expected Behavior

The user receives the complete response — either because the finalize edit carries the completed text, or because a mismatch between the finalized text and the completed response prevents `content_delivered=True` and lets the normal final send deliver the remainder.

### Actual Behavior

Complete response 272 characters, delivered 155, ending mid-sentence. No retry, no continuation, no warning.

```text
01:18:05,977 DEBUG ExtBot: Calling `editMessageText` {'text': '<151 chars><cursor ▉>', ...}
01:18:06,607 INFO  run_agent: OpenAI client closed (stream_request_complete, ...)
01:18:06,609 INFO  agent.conversation_loop: API call #1: ... out=274 total=43181 latency=5.3s
[...]

PR #98720, opened 2026-08-30
fix(telegram): report the clipped stream preview instead of a bare success (#98552)

### LINKED ISSUE #98552: Telegram streaming finalize fails — message truncated well under 4096 chars, content_delivered=True false positive
## Bug Description

When Telegram streaming is enabled, a **624-character message** (well under the 4096 UTF-16 limit) was truncated mid-delivery. The streaming cursor `▉` remained visible at the truncation point, and the gateway incorrectly marked the delivery as complete.

**User-visible result:**
```
- 已发布: 713 条（08-26 时 578，过去 4 天又 ▉
```

**Expected (full content):**
```
- 已发布: 713 条（08-26 时 578，过去 4 天又发了 135 条）
```

## Evidence from Logs

```
2026-08-30 12:04:46,661 INFO hermes_plugins.telegram_platform.adapter: [Telegram] Flushing text batch agent:main:telegram:group:-1003900203906:14321 (333 chars)
2026-08-30 12:05:11,006 INFO gateway.run: Suppressing normal final send for session agent:main:telegram:group:-1003900203906:14321: final delivery already confirmed (streamed=True previewed=False content_delivered=True).
```

Key observations:
1. Only **333 chars** were flushed to Telegram (out of 624 total)
2. The gateway logged `content_delivered=True` — falsely confirming delivery was complete
3. The `▉` streaming cursor remained in the final message (finalize did not strip it)
4. No overflow split or continuation message was attempted

## Root Cause Analysis

The finalize path in `gateway/stream_consumer.py` appears to have a race condition or incorrect state tracking:

1. The streaming edit was flushed (333 chars) but the full content was never sent
2. `_visible_prefix()` may have matched the truncated text against `final_text`, causing the system to believe content was "already shown"
3. The `_try_strip_cursor()` path ran but failed to produce a clean message (or the edit itself was truncated)
4. `_final_response_sent` was set to `True` prematurely, suppressing the normal final send

This is **not** a 4096-character overflow issue. The message was 624 chars total. The bug is in the finalize completion detection logic.

## Environment

- Hermes Agent: latest main (2026-08-30)
- Platform: Telegram group chat with topics
- Streaming config: `telegram.streaming: true`, edit-based transport
- Model: mimo-v2.5 (xiaomi provider)
- OS: Linux

## Reproduction

Not consistently reproducible — appears to be a timing/state issue during streaming. The bug manifested during a normal assistant response with tool calls (file reads + verification).

## Workaround

Set `telegram.streaming: false` in config.yaml to disable streaming entirely.

```bash
hermes config set telegram.streaming false
```

## Related Issues

- #51961 (closed as cannot-reproduce — 4096 overflow scenario, different root cause)
- #29200 (truncation at unclosed backtick during finalize)
- #51828 (truncation triggers full re-generation)

JoaoMarcos44:
## Problem
Fixes #98552.

`gateway/run.py` suppresses its own final send when the stream consumer claims the turn final already reached the user:

```text
Suppressing normal final send for session ...: final delivery already confirmed (streamed=True previewed=False content_delivered=True).
```

That claim is only as good as the consumer's model of what is on screen — and on Telegram that model can be wrong by design.

Telegram refuses to **split** an oversized *progressive* edit: splitting mid-stream moves the editable message id, so the next accumulated-token edit re-splits and the reply duplicates (#48648). The adapter therefore **clips** the payload to one message and edits with the clipped text. The call succeeded, so `edit_message()` returned a bare `SendResult(success=True)` — with no way for the caller to learn that Telegram stored *less* than it was handed.

`GatewayStreamConsumer._send_or_edit()` then advanced `_last_sent_text` from the text it **sent**. Every "what has the user already seen" derivation is built on that one field:

| Derivation | What it is used for | Effect when `_last_sent_text` is too long |
|---|---|---|
| `_visible_prefix()` | on-screen text | over-reports the preview |
| `_continuation_text()` | the tail a fallback send still owes | fallback starts past the visible end — the middle of the answer is dropped |
| `_mark_skip_redundant_finalize()` / failed-final-edit guard | turn-final payload the gateway reconciles | a clipped preview reconciles as the completed `final_response` |

So `delivered_final_matches()` returns `True` for a message that only holds the first 4,096 units, the gateway suppresses its corrective send, and the user is left with a frozen preview and a stuck streaming cursor. This is the same failure family as #71643 / #78541 / #82656, and it is the one member of it whose "ACK for text the platform never stored" is produced **inside Hermes**, not by the network.

This path is not exotic: `TelegramAdapter.streaming_overflow_limit()` deliberately raises the consumer's own split threshold to the 32,768 rich-message cap, so any streamed reply over 4,096 characters on a rich-capable bot relies on the adapter's clip.

```mermaid
graph TD
    A[Streamed Deltas] -->|accumulated over 4096| B[Progressive Edit]
    B -->|clip to one message| C[Telegram Stores 4096]
    C -->|bare SendResult success=true| D[Consumer Records Full Text]
    D --> E[visible_prefix over-reports]
    D --> F[continuation_text skips the middle]
    D --> G[turn-final payload is un-stored text]
    G --> H[delivered_final_matches returns True]
    H --> I[Gateway Suppresses Final Send]
    I --> J[Frozen Preview and Stuck Cursor]
```

## Solution

Make the ACK honest, then let the existing recovery machinery do its job.

1. **`plugins/platforms/telegram/adapter.py`** — all three clipped-preview returns (pre-flight clip, saturated-preview dedup, and the reactive `message_too_long` retry) now go through `_stream_preview_truncated_result()`, returning `raw_response={"stream_preview_truncated": True, "delivered_prefix": <stored text>}`. `delivered_prefix` is the field the existing `partial_overflow` contract already carries on the failure branch, so no new vocabulary is introduced.
2. **`gateway/stream_consumer.py`** — the successful-edit path sets `_last_sent_text` through the new `_delivered_text_for()` helper, tracking what the platform **stored** rather than what was **requested**. Adapters that deliver exactly what they were given report nothing and keep the previous behaviour.

Nothing downstream changed: once `_visible_prefix()` is accurate, `_continuation_text()` computes the real missing tail and `delivered_final_matches()` reports the mismatch, so `gateway/run.py` takes its existing stale-finalize reconciliation / normal final-send path.

```mermaid
%%{init: {'theme': 'dark', 'themeVariables': { 'primaryColor': '#00f0ff', 'mainBkg': '#0a0a16', 'primaryTextColor': '#ffffff', 'primaryBorderColor': '#ff007f', 'lineColor': '#00f0ff'}}}%%
graph TD
    A[Progressive Edit] -->|clip to one message| B[Telegram Stores 4096]
    B -->|delivered_prefix reported| C[Consumer Tracks Stored Text]
    C --> D[visible_prefix matches the screen]
    D --> E[Fallback Sends the Real Tail]
    D --> F[delivered_final_matches returns False]
    F --> G[Gateway Delivers the Complete Answer]
```

## Notes on the report

Two details in #98552 did not survive verification, and they are worth recording so the next reader does not chase them:

- The quoted `[Telegram] Flushing text batch ... (333 chars)` line is **inbound** message aggregation (`TelegramAdapter._flush_text_batch` calling `handle_message`), i.e. the user's own message being debounced before dispatch. It is not the outbound reply, so "only 333 of 624 chars were flushed" is not what that log shows.
- The proposed cause (an interrupted stream leaving no record, `delivered_final_matches()` returning `None`, and the gateway keeping legacy trust) is not reachable from the consumer's own bookkeeping. Driving the real `GatewayStreamConsumer` against a Telegram-shaped adapter across 2,048 honest-transport scenarios (flood / timeout / permanent-error scripts on edit and send, tool boundaries, authoritative `finish(final_text)` adoption), 600 delta/boundary/commentary schedules, and 729 overflow-split scenarios produced **zero** states where the consumer claimed delivery it could not back up. The in-tree contract matrix (`tests/gateway/test_suppression_contract_matrix.py`) reaches the same conclusion for honest transports.

What *is* reachable is the clipped-preview ACK above, which is exactly the "edit ACKed but never rendered" residue that matrix documents as its single `xfail`.

## Verification

```bash
pytest tests/gateway/test_truncated_preview_ack.py \
       tests/gateway/test_suppression_contract_matrix.py \
       tests/gateway/test_stale_finalize_suppression.py \
       tests/gateway/test_stream_final_contract.py \
       tests/gateway/test_telegram_format.py \
       tests/gateway/test_stream_consumer.py \
       tests/gateway/test_telegram_final_delivery.py \
       tests/gateway/test_split_final_suffix_reconcile.py
```

`185 passed, 1 xfailed`. The six new tests in `tests/gateway/test_truncated_preview_ack.py` were confirmed non-vacuous: five of them fail on the unpatched tree, including the end-to-end one that reproduces the loss (`rendered_len=4117` vs `final_len=5226`, with `verdict=True` and both suppression flags set).

Also ran `tests/gateway/test_stream_consumer_draft.py`, `test_stream_consumer_fresh_final.py`, `test_stream_consumer_wecom_native.py`, `test_duplicate_reply_suppression.py`, `test_stream_consumer_tool_progress.py`, `test_stream_final_adoption_gate.py` and `test_discord_edit_message_overflow.py` — all green. One unrelated pre-existing failure on this machine (`test_update_streaming.py::TestUpdateCommandGatewayFlag::test_spawns_with_gateway_flag`) reproduces on a clean tree.

## Relationship to #98583

#98583 targets the same issue from the reconciliation side, changing `delivered_final_matches()` to compare `_visible_prefix()` against the target when `_delivered_final_text is None`. Its regression test reaches that state by subclassing `GatewayStreamConsumer.run()` and assigning the flags after the run returns — the state the fuzzing above could not produce from the real consumer. The two changes do not conflict; this one removes the source of the bad prefix rather than reinterpreting it.

### DIFF
--- a/gateway/stream_consumer.py
+++ b/gateway/stream_consumer.py
@@ -2080,6 +2080,24 @@ class GatewayStreamConsumer:
             logger.error("Stream send chunk error: %s", e)
             return reply_to_id
 
+    @staticmethod
+    def _delivered_text_for(result: Any, sent_text: str) -> str:
+        """Return the text the adapter reports the platform actually stored.
+
+        Adapters that clip an oversized progressive edit instead of splitting
+        it (Telegram's saturated stream preview) still report success, and
+        surface what landed as ``raw_response["delivered_prefix"]`` — the same
+        field the ``partial_overflow`` contract uses on the failure branch.
+        Adapters that deliver exactly what they were given report nothing and
+        keep ``sent_text``.
+        """
+        raw = getattr(result, "raw_response", None)
+        if isinstance(raw, dict):
+            delivered = raw.get("delivered_prefix")
+            if isinstance(delivered, str) and delivered:
+                return delivered
+        return sent_text
+
     def _visible_prefix(self) -> str:
         """Return the visible text already shown in the streamed message."""
         prefix = self._last_sent_text or ""
@@ -3362,7 +3380,17 @@ class GatewayStreamConsumer:
                             self._last_sent_text = ""
                             self._notify_new_message()
                         else:
-                            self._last_sent_text = text
+                            # Track what the platform actually STORED, not what
+                            # we asked it to store.  An adapter may accept an
+                            # oversized progressive edit and keep only a
+                            # truncated preview (Telegram, #48648); advancing
+                            # from ``text`` then makes every "already seen"
+                            # derivation — _visible_prefix, _continuation_text,
+                            # the turn-final payload record — describe text no
+                            # API call ever carried (#98552).
+                            self._last_sent_text = self._delivered_text_for(
+                                result, text,
+                            )
                         # Successful edit — reset flood strike counter
                         self._flood_strikes = 0
                         return True
--- a/plugins/platforms/telegram/adapter.py
+++ b/plugins/platforms/telegram/adapter.py
@@ -230,6 +230,33 @@ _TELEGRAM_IMAGE_EXTENSIONS = {".png", ".jpg", ".jpeg", ".webp", ".gif"}
 _FLOOD_INLINE_WAIT_CAP_SECS = 5.0
 
 
+def _stream_preview_truncated_result(
+    message_id: str, delivered_prefix: str
+) -> "SendResult":
+    """Success result for a mid-stream edit Telegram stored only in part.
+
+    A progressive edit that overflows the 4,096 UTF-16 cap is truncated to one
+    message instead of being split (#48648), so the text that reaches Telegram
+    is SHORTER than the text the caller passed.  Reporting a bare success made
+    that invisible: the stream consumer advanced ``_last_sent_text`` — and
+    every "what has the user already seen" derivation built on it
+    (``_visible_prefix``, ``_continuation_text``, the turn-final payload
+    record) — to text no API call ever carried, so a later fallback skipped
+    the un-stored middle and the gateway suppressed its corrective send
+    (#98552).  ``delivered_prefix`` states what Telegram actually holds; the
+    existing ``partial_overflow`` contract carries the same field on the
+    failure branch.
+    """
+    return SendResult(
+        success=True,
+        message_id=message_id,
+        raw_response={
+            "stream_preview_truncated": True,
+            "delivered_prefix": delivered_prefix,
+        },
+    )
+
+
 def _flood_cap_result(wait: float) -> "SendResult":
     """The shared fail-closed SendResult for an over-cap flood wait."""
     return SendResult(
@@ -5744,7 +5771,7 @@ class TelegramAdapter(BasePlatformAdapter):
             # stream trips flood control (200s+ penalties) and hangs the final
             # delivery. Skip silently until finalize.
             if self._last_overflow_preview.get(_preview_key) == content:
-                return SendResult(success=True, message_id=message_id)
+                return _stream_preview_truncated_result(message_id, content)
         elif not finalize:
             # Content shrank back under the cap (segment break / new message
             # id) — clear stale saturation state so dedup can't mask a real
@@ -5760,6 +5787,7 @@ class TelegramAdapter(BasePlatformAdapter):
                 )
                 if _saturated_preview:
                     self._last_overflow_preview[_preview_key] = content
+                    return _stream_preview_truncated_result(message_id, content)
                 return SendResult(success=True, message_id=message_id)
 
             formatted = self.format_message(content)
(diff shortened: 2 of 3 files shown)


Enough1122:

**Overall:** Fixes stream progress tracking when Telegram truncates oversized progressive edits instead of splitting (#98552, #48648).

**Correctness:**
- `plugins/platforms/telegram/adapter.py:230` — new `_stream_preview_truncated_result(message_id, delivered_prefix)` returns `success=True` with `raw_response={"stream_preview_truncated":True,"delivered_prefix":...}`.
- `gateway/stream_consumer.py:2080` — `_delivered_text_for(result, sent_text)` reads `delivered_prefix` if present; ` _send_or_edit` sets `self._last_sent_text = _delivered_text_for(result,text)` instead of `text`. Ensures `_visible_prefix` / `_continuation_text` / payload record reflect what Telegram actually stored, preventing skipped middle segment.
- Fallback deduplication `(_last_overflow_preview==content)` now returns truncated result instead of bare success.

**Non-blocking:**
- `delivered_prefix` contract reused from `partial_overflow` failure branch; consistent.

**Verdict:** LGTM.

PR #98721, opened 2026-08-30
fix(config): prove env-template provenance via runtime secret authorities

### LINKED ISSUE #98717: security(config): ${VAR} template can be replaced by a live secret when the env var is absent at save time
## Summary

`save_config()`'s env-template preservation guard can still persist a live credential as plaintext when the raw file contains a `${VAR}` template but that variable is **not present in `os.environ` at save time** and the resolved credential reached the config object through another runtime authority.

This is a narrower hole in the existing preservation mechanism, not a re-report of the ordinary `load_config()` expansion case.

## Current-main source-of-truth

Verified on `main@4f22543509d1b91dc45bcb369447126c5eb14fb7`.

`hermes_cli/config.py::_preserve_env_ref_templates()` preserves a raw `${VAR}` only when `current` matches either:

1. an expansion that can be recomputed from the raw template using the process environment, or
2. the previous `loaded_expanded` value.

If `${VAR}` is unresolved because `VAR` is absent from `os.environ`, but a qualified runtime source has already supplied the live credential into `current`, neither comparison can prove provenance and the helper falls through to the caller-provided literal. A later `save_config()` can therefore replace the on-disk template with plaintext.

## Concrete prior reproduction / provenance

Prior work by @dizhaky in closed-unmerged #98029 identified and directly reproduced this exact boundary in a downstream environment:

> var UNSET, token came from elsewhere -> template preserved? False
> var SET and matches -> template preserved? True

That PR was closed immediately because it was accidentally opened against this upstream repository and was intended for a downstream fork. **Do not treat #98029's 7,656-file branch topology as an upstream implementation carrier.** Preserve the defect provenance only.

The reported production shape was a raw `Bearer ${UNIFIED_GATEWAY_BEARER_TOKEN}` template whose live bearer value was available to the daemon but not reproducible through `os.environ` at the save point. The literal then replaced the template on disk.

## Why the existing guard does not settle this class

#11892 established the save-chokepoint template-preservation defense. #94918's investigation correctly showed that ordinary round-trips are already protected when the current value equals the process-env expansion or the previous expanded snapshot; #94918 ultimately traced its concrete leak to a different path.

This case is the other side of that proof: **the secret value is real, but the save path cannot reconstruct its origin from `os.environ` or the prior expanded snapshot.** Value equality is therefore insufficient as the only provenance signal.

## Minimal regression shape

A regression should cover the helper and a real `save_config()` path with a disposable `HERMES_HOME`:

```python
raw = {"gateway": {"authorization": "Bearer ${TEST_BEARER_TOKEN}"}}
current = {"gateway": {"authorization": "Bearer opaque-live-secret-value"}}

# TEST_BEARER_TOKEN deliberately absent from os.environ at save time.
# The live value is supplied through the runtime/config authority under test,
# not by expanding the raw template here.
```

**Expected:** unrelated persistence cannot replace the raw `${TEST_BEARER_TOKEN}` template with the live secret.

**Rejected fix shape:** blindly preserving every unresolved credential-looking `${VAR}` template. That can erase an intentional caller-owned replacement. The regression matrix must include an explicit intentional-edit positive control.

## Acceptance contract

- Preserve the raw env-ref template across unrelated writes even when the live credential came from a non-`os.environ` runtime source.
- Preserve intentional explicit replacement semantics; a user-owned edit must still be able to replace the template.
- Bind the decision to write provenance/intent, not only secret-looking strings or ambient value equality.
- Exercise the real persistence path in isolated state and read the resulting `config.yaml` back byte-for-byte.
- Cover rotation/load-save asymmetry, empty/clear semantics, ordinary non-secret `${VAR}` values, and the intentional-replacement positive control.
- Keep #98029/@dizhaky provenance explicit without importing its wrong-repository branch history.

Related: #11864, #94918, #11892, #98029.

liuhao1024:
## What does this PR do?

`save_config()`'s env-template preservation guard proved that a live value originated from the raw `${VAR}` template only when `os.environ` could recompute the expansion. When the variable is absent from `os.environ` at save time but the resolved credential reached the config object through another runtime authority, none of the three existing comparisons could prove provenance and the helper fell through to the caller-provided literal — persisting the live secret as plaintext in place of the on-disk template (#98717).

This PR extends the provenance proof with a save-side expansion that resolves each `${VAR}` / `${env:VAR}` name through the runtime secret authorities, in order:

1. `os.environ` (unchanged, fast path),
2. the scope-aware `agent.secret_scope.get_secret` — under a multiplexed gateway turn the profile scope supplies a live credential that never appears in `os.environ`,
3. `~/.hermes/.env` via `load_env()`.

A value any of those authorities can recompute from the raw template keeps the template on disk. A caller-typed literal that no authority can recompute remains an intentional replacement, so replacement semantics are unchanged — the decision is bound to recomputable provenance, not to secret-looking strings or ambient equality. Defect provenance for this boundary: @dizhaky's closed-unmerged #98029 reproduction ("var UNSET, token came from elsewhere -> template preserved? False").

## Related Issue

Fixes #98717

## Type of Change

- [x] 🔒 Security fix

## Changes Made

- `hermes_cli/config.py`: added `_runtime_env_ref_value()` (resolve one env-ref name via os.environ → `get_secret` → `.env`, exceptions treated as a miss) and `_expand_env_refs_via_runtime()` (same `${VAR}`/`${env:VAR}` grammar as `_env_expand_match`, unresolved/non-env refs stay verbatim).
- `hermes_cli/config.py`: `_preserve_env_ref_templates()` gains a fourth preservation rule after the existing environment-expansion check: runtime-authority expansion equality → keep the raw template.
- `tests/hermes_cli/test_config_env_refs.py`: regression matrix — scope-supplied live secret preserved, dotenv-supplied value preserved, intentional replacement under scope still lands as literal (positive control), non-secret `${VAR}` template preserved, clearing a templated value stays cleared.

## How to Test

1. `python -m pytest tests/hermes_cli/test_config_env_refs.py -v` — Observed result: 7 passed (2 pre-existing + 5 new).
2. `python -m pytest tests/hermes_cli/test_config_env_expansion.py tests/hermes_cli/test_config_env_ref_parity.py tests/hermes_cli/test_config_loader_e2e.py tests/hermes_cli/test_config_set_coercion.py tests/hermes_cli/test_config_set_list_values.py tests/hermes_cli/test_config_validation.py tests/hermes_cli/test_config_read_guard.py` — Observed result: 49 passed.
3. Wider save_config surface (`tests/test_secret_scope_plugin_families.py`, `tests/hermes_cli/test_model_switch_custom_providers.py`, `tests/hermes_cli/test_setup.py` and others) — Observed result: 179 passed.
4. `python -m ruff check hermes_cli/config.py tests/hermes_cli/test_config_env_refs.py` and `python scripts/check-windows-footguns.py hermes_cli/config.py` — Observed result: all clean.
5. The fail-closed regression mirrors the issue's minimal shape: raw `authorization: Bearer ${TEST_BEARER_TOKEN}`, variable absent from `os.environ`, live value supplied through the scope authority, unrelated setting persisted → the written `config.yaml` still contains `Bearer ${TEST_BEARER_TOKEN}` byte-for-byte and never the live secret.

### Code

- [x] I've updated relevant documentation (README, `docs/`, docstrings) — docstrings updated on the guard and new helpers
- [x] I've considered cross-platform impact (Windows, macOS) per the [compatibility guide](https://github.com/NousResearch/hermes-agent/blob/main/CONTRIBUTING.md#cross-platform-compatibility) — pure-Python env/dict logic, no OS-specific calls

### DIFF
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -2785,6 +2785,53 @@ def _env_ref_var_name(ref: str) -> Optional[str]:
     return ref
 
 
+def _runtime_env_ref_value(name: str) -> Optional[str]:
+    """Resolve one env-ref name through the runtime secret authorities.
+
+    ``os.environ`` first (same as ``_env_expand_match``), then the scope-aware
+    ``agent.secret_scope.get_secret`` and ``~/.hermes/.env``. Those two can
+    supply a live credential that never passes through ``os.environ`` — e.g.
+    a value resolved under an installed profile scope during a multiplexed
+    gateway turn, or one that only lives in the dotenv file. Used solely by
+    the save-side template-preservation proof; load-side expansion stays
+    environment-only (#98717).
+    """
+    val = os.environ.get(name)
+    if val is not None:
+        return val
+    try:
+        from agent.secret_scope import get_secret
+
+        val = get_secret(name)
+    except Exception:
+        val = None
+    if val is not None:
+        return val
+    try:
+        return load_env().get(name)
+    except Exception:
+        return None
+
+
+def _expand_env_refs_via_runtime(raw: str) -> str:
+    """Expand ``${VAR}`` / ``${env:VAR}`` in one string via runtime authorities.
+
+    Mirrors ``_env_expand_match``'s ref grammar but resolves each name with
+    ``_runtime_env_ref_value``; unresolved refs stay verbatim. Non-env
+    SecretRef sources (``file:``, ``vault:``, ...) stay verbatim exactly as
+    the load-side expander leaves them.
+    """
+
+    def _match(m: "re.Match[str]") -> str:
+        name = _env_ref_var_name(m.group(1))
+        if name is None:
+            return m.group(0)
+        val = _runtime_env_ref_value(name)
+        return val if val is not None else m.group(0)
+
+    return re.sub(r"\${([^}]+)}", _match, raw)
+
+
 def _expand_env_vars(obj):
     """Recursively expand ``${VAR}`` / ``${env:VAR}`` references in config
     values.
@@ -2861,6 +2908,14 @@ def _preserve_env_ref_templates(current, raw, loaded_expanded=None):
     the current environment expansion of ``raw``. This handles env-var
     rotation between load and save while still treating mixed literal/template
     string edits as caller-owned once their rendered value diverges.
+
+    The provenance proof also consults the runtime secret authorities
+    (profile scope via ``get_secret``, then ``~/.hermes/.env``): a live
+    credential can reach the config object through those without ever
+    appearing in ``os.environ``, and value equality against the bare
+    environment alone cannot prove the template is its origin (#98717).
+    A caller-typed literal that no authority can recompute stays
+    caller-owned, so intentional replacement semantics are unchanged.
     """
     if isinstance(current, str) and isinstance(raw, str) and re.search(r"\${[^}]+}", raw):
         if current == raw:
@@ -2868,6 +2923,8 @@ def _preserve_env_ref_templates(current, raw, loaded_expanded=None):
         if isinstance(loaded_expanded, str) and current == loaded_expanded:
             return raw
         if _expand_env_vars(raw) == current:
+            return raw
+        if _expand_env_refs_via_runtime(raw) == current:
             return raw
         return current
 
--- a/tests/hermes_cli/test_config_env_refs.py
+++ b/tests/hermes_cli/test_config_env_refs.py
@@ -59,6 +59,146 @@ def test_save_config_allows_intentional_secret_value_change(monkeypatch, tmp_pat
     assert "${TU_ZI_API_KEY}" not in saved
 
 
+# --- Runtime-authority provenance proof (#98717) ---------------------------
+#
+# A live credential can reach the config object through a runtime authority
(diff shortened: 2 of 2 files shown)


andrexibiza (commented):
Exact-head source and hosted execution are coherent on `48ecd5d90f1dbc6ac62345147361c7a06ef663ce`: this is a one-commit/two-file train, and CI 33323989258, Docker 33323988874, and Nix 33323988866 all succeeded on this exact object. The provenance direction also closes the reported value-equality gap without collapsing intentional replacement: scope-provided and dotenv-provided values can re-prove the raw env ref, while an unrelated caller literal still lands as a literal.

There is one placement blocker before this can clear acceptance: the implementation extends `hermes_cli/config.py`, which is already well above the 2,000-line ceiling (the touched persistence/env-ref region itself is beyond line 2,700). This workstream does not add behavior in-place to a godfile. Extract the env-ref/save-persistence ownership slice so every modified file — including `hermes_cli/config.py` after extraction — is below 2,000 lines, preserve the same real `save_config()` regression matrix and intentional-edit positive control, then reacquire exact-head CI/Docker/Nix on the rematerialized object. Keep the #98029/@dizhaky provenance edge intact. No semantic objection remains to the runtime-authority provenance mechanism itself on this head.

PR #98726, opened 2026-08-30
fix(bedrock): correct Llama 4 context windows (128K -> measured 1M/3.5M), add six measured rows

JiaDe-Wu:
## The bug

`BEDROCK_CONTEXT_LENGTHS` lists both Llama 4 models at `128_000`. Neither is 128K. Measured against Converse, **Maverick enforces 1,048,576** and **Scout enforces 3,500,000** — the table was capping them at 1/8 and 1/27 of the real window.

These numbers feed `get_bedrock_context_length()`, which feeds the compaction threshold. The user-visible effect is that a Llama 4 session starts summarising away context almost immediately and keeps doing it, while the model could have held the whole conversation. Silent, and it looks like the agent is just forgetful.

Four more models were absent from the table entirely and silently inherited the 128K default:

| model | table before | measured | ratio |
|---|---|---|---|
| `meta.llama4-maverick` | 128,000 | **1,048,576** | 8.2× under |
| `meta.llama4-scout` | 128,000 | **3,500,000** | 27.3× under |
| `mistral.pixtral-large` | *(default)* 128,000 | **131,072** | 1.02× |
| `deepseek.v3.2` | *(default)* 128,000 | **163,840** | 1.28× |
| `minimax.minimax-m2` | *(default)* 128,000 | **196,608** | 1.54× |
| `kimi-k2` | *(default)* 128,000 | **262,144** | 2.05× |

## How the numbers were obtained

Send Converse a prompt padded just past the suspected window and read the maximum out of the resulting `ValidationException`. Bedrock validates request length *before* inference, so a rejected probe is not billed — which is what makes this cheap enough to do exhaustively.

Verbatim, for Scout:

```
An error occurred (ValidationException) when calling the Converse operation:
The model returned the following errors: This model's maximum context length
is 3500000 tokens. Please reduce the length of the prompt
```

Cross-checks, because a single probe proves very little:

- **Maverick rejects at the same 1,048,576 in `us-east-1` and `us-east-2`** — not a regional quota artefact.
- **MiniMax is identical on `minimax-m2`, `minimax-m2.1` and `minimax-m2.5`** (196,608), so one key covers the family.
- **Kimi K2 is identical on `moonshot.kimi-k2-thinking` and `moonshotai.kimi-k2.5`** (262,144).
- **Negative results were kept too.** Nova Pro rejected at 345,000 and DeepSeek R1 at 147,200, both consistent with their existing rows — so I changed nothing there. Only rows I could show were wrong were touched.

## Two judgement calls I want to flag explicitly

**1. Scout is not 10M on Bedrock, whatever its model card says.**

The [Bedrock model card for Llama 4 Scout](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-meta-llama-4-scout-17b-instruct.html) states 10M. The API rejects at exactly 3,500,000. The card describes Meta's native model; Bedrock serves less. I went with the measurement, because trusting the doc here would overshoot by ~3× and that failure mode is strictly worse than the one being fixed — it converts *premature compaction* (annoying) into *hard `ValidationException`s* (broken). I had 3.5M in mind before measuring, then found the doc contradicting me, and did not write either number down until a probe settled it.

**2. Maverick's card rounds to "1M"; I used 1,048,576.**

The API reports the exact binary multiple, so that is what it enforces. Writing 1,000,000 would be safe but leaves 48K on the table for no reason.

## The Kimi key deliberately has no vendor prefix

Bedrock ships this family as both `moonshot.kimi-k2-thinking` and `moonshotai.kimi-k2.5`. `_static_bedrock_context_length()` does longest-substring matching, so the prefix-less key `"kimi-k2"` covers both without duplicating the row. That is load-bearing and easy to "tidy up" into a break, hence the comment on it.

Relatedly: **every key in this change was checked against `list-foundation-models` output in `us-east-1`/`us-east-2`/`us-west-2`**, because a table key that matches no real model ID is a silent no-op that a unit test written against the key itself will happily pass. `deepseek.v3.2`, `minimax.minimax-m2`, `moonshot.kimi-k2-thinking` and `moonshotai.kimi-k2.5` are literal Bedrock model IDs with no `-v1:0` suffix; the tests use the real IDs.

## A parser note, no behaviour change

Several Bedrock vendors report a length error with **no number in it at all**:

- Nova: `Input Tokens Exceeded: Number of input tokens exceeds maximum length.`
- DeepSeek R1: `Input is too long for requested model.`

`parse_context_limit_from_error()` returns `None` for these, which is correct — and it is exactly *why* this static table is those models' only source of truth, since `probe_bedrock_context_length()` can never learn their window. I added that to the docstring so the next person doesn't read the `None` as a parser gap and "fix" it into a guess. **The parser code is untouched.**

I did briefly raise the parser's sanity ceiling from `10_000_000` to `10_485_760`, on the theory that a 10Mi Scout limit would be discarded. Measuring Scout at 3,500,000 disproved the premise, so I reverted it rather than ship a change whose only justification I had just refuted.

## Tests

29 new tests in `tests/agent/test_bedrock_adapter.py`:

- each measured window resolves from the **real** model ID, including the `us.` inference-profile forms the picker actually offers;
- the regression form — none of these models may resolve to `BEDROCK_DEFAULT_CONTEXT_LENGTH`;
- both shadowing hazards: `deepseek.v3.2` must beat the `deepseek.v3` substring, and the Scout/Maverick rows must not collapse into each other;
- Llama 3 stays at 128K (only Llama 4 is long-context);
- the verbatim captured error dialects against `parse_context_limit_from_error`, including the two numberless ones.

**Mutation check:** reverting only the table, keeping the tests, turns **24 of the 29 red**. The 5 that stay green are the Llama-3 row and the four parser-dialect tests — which is right, since those document existing behaviour rather than guarding this change.

`ruff` clean. `tests/agent/test_bedrock_adapter.py` 105 passed; the three `model_metadata` suites 150 passed.

### DIFF
--- a/agent/bedrock_adapter.py
+++ b/agent/bedrock_adapter.py
@@ -1887,6 +1887,15 @@ def classify_bedrock_error(error_message: str) -> str:
 # Static fallback table for models where the Bedrock API doesn't expose
 # context window sizes.  Used by agent/model_metadata.py when dynamic
 # detection is unavailable.
+#
+# Rows marked "measured" were obtained by sending Converse a prompt padded
+# just past the suspected window and reading the maximum out of the resulting
+# ValidationException.  Length validation runs before inference, so a rejected
+# probe costs nothing.  This matters because several vendors return a length
+# error with no number in it at all — Nova ("Input Tokens Exceeded: Number of
+# input tokens exceeds maximum length.") and DeepSeek R1 ("Input is too long
+# for requested model.") both do — which means probe_bedrock_context_length()
+# returns None for them and this table is their only source of truth.
 
 BEDROCK_CONTEXT_LENGTHS: Dict[str, int] = {
     # Anthropic Claude models on Bedrock.
@@ -1921,13 +1930,36 @@ BEDROCK_CONTEXT_LENGTHS: Dict[str, int] = {
     "amazon.nova-lite":              300_000,
     "amazon.nova-micro":             128_000,
     # Meta Llama
-    "meta.llama4-maverick":          128_000,
-    "meta.llama4-scout":             128_000,
+    # Llama 4 is long-context, not 128K. Both rows were capping it at an
+    # eighth (Maverick) and a twenty-seventh (Scout) of the real window, so a
+    # Llama 4 session compacted almost immediately.
+    # Maverick measured against Converse in us-east-1 and us-east-2 — both
+    # reject at the same value: "This model's maximum context length is
+    # 1048576 tokens". The model card rounds that to "1M"; the API enforces
+    # the exact binary multiple, so use the enforced number.
+    # https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-meta-llama-4-maverick-17b-instruct.html
+    "meta.llama4-maverick":          1_048_576,
+    # Scout measured at 3,500,000: "This model's maximum context length is
+    # 3500000 tokens". Deliberately *not* the 10M its Bedrock model card
+    # states — that is Meta's native window, and Bedrock serves less. Taking
+    # the documented figure here would overshoot by 3x and turn premature
+    # compaction into hard ValidationExceptions, so the measured value wins.
+    # https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-meta-llama-4-scout-17b-instruct.html
+    "meta.llama4-scout":             3_500_000,
     "meta.llama3-3-70b-instruct":    128_000,
     # Mistral
     "mistral.mistral-large":         128_000,
+    "mistral.pixtral-large":         131_072,   # measured
     # DeepSeek
     "deepseek.v3":                   128_000,
+    "deepseek.v3.2":                 163_840,   # measured; longest-match beats "deepseek.v3"
+    # MiniMax — measured identically on minimax-m2, m2.1 and m2.5.
+    "minimax.minimax-m2":            196_608,
+    # Moonshot Kimi K2. Bedrock ships this family under two different vendor
+    # prefixes ("moonshot." and "moonshotai."), so the key deliberately omits
+    # the prefix to match both. Measured identically on kimi-k2-thinking and
+    # kimi-k2.5.
+    "kimi-k2":                       262_144,
     # OpenAI on Bedrock (Mantle/Responses route)
     # https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards-openai.html
     "openai.gpt-5.5":                272_000,
--- a/agent/model_metadata.py
+++ b/agent/model_metadata.py
@@ -1709,6 +1709,13 @@ def parse_context_limit_from_error(error_msg: str) -> Optional[int]:
       - "Maximum context size 32768 exceeded"
       - "model's max context length is 65536"
       - "input token count is 32825 but model only supports up to 32768"
+      - "This model's maximum context length is 1048576 tokens" (Bedrock Llama 4)
+
+    Some do not, and there is nothing to extract: Bedrock's Nova models say
+    "Input Tokens Exceeded: Number of input tokens exceeds maximum length."
+    and DeepSeek R1 says "Input is too long for requested model.", neither of
+    which names a number.  ``None`` is the correct answer there — callers fall
+    back to their static table rather than inventing a window.
     """
     error_lower = error_msg.lower()
     # Pattern: look for numbers near context-related keywords
--- a/tests/agent/test_bedrock_adapter.py
+++ b/tests/agent/test_bedrock_adapter.py
@@ -1112,6 +1112,119 @@ class TestBedrockContextProbe:
                 region="eu-central-1") == 1_000_000
 
 
+class TestMeasuredBedrockContextWindows:
+    """Pin context windows that were measured against live Bedrock Converse.
+
+    Each value here came from padding a Converse request just past the
+    suspected window and reading the maximum out of the ValidationException.
+    The tests assert against the real model IDs Bedrock serves — including the
+    ``us.`` inference-profile forms the picker actually offers — because the
+    lookup is substring-based and it is the *resolution* that regressed, not
+    the literal dict entry.
+    """
+
+    # (model_id, expected_window)
+    MEASURED = [
+        # Llama 4 was the actual bug: both rows read 128_000, capping Maverick
+        # at 1/8 and Scout at 1/27 of their real windows.
+        ("meta.llama4-maverick-17b-instruct-v1:0", 1_048_576),
+        ("us.meta.llama4-maverick-17b-instruct-v1:0", 1_048_576),
+        ("meta.llama4-scout-17b-instruct-v1:0", 3_500_000),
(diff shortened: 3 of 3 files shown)


Enough1122:

**Overall:** Corrects Bedrock context-window fallback table from live probing (Maverick 128K→1,048,576, Scout 128K→3,500,000, etc.) and documents `None` fallback for non-numeric errors.

**Correctness:**
- `agent/bedrock_adapter.py:1800` — updates `BEDROCK_CONTEXT_LENGTHS` with measured values (Maverick 1,048,576 exact binary, Scout 3,500,000 not 10M card, `pixtral-large` 131072, `deepseek.v3.2` 163840, `minimax-m2` 196608, `kimi-k2` 262144). Chooses measured over documented when Bedrock serves less.
- `agent/model_metadata.py:1679` — docstring notes Nova/DeepSeek error msgs contain no number → `parse_context_limit_from_error` correctly returns `None`.

**Non-blocking:**
- Scout 10M→3.5M is intentional conservative; document as above.

**Verdict:** LGTM.

JiaDe-Wu:
Rebased onto current `main` (`00b2e03c`) — this PR had gone `CONFLICTING`.

The only conflict was in the `parse_context_limit_from_error` docstring, where `main` had added a bullet for the Gemini *"input token count is 32825 but model only supports up to 32768"* shape while this branch added one for the Bedrock Llama 4 shape. Both are additive documentation of real error strings, so I kept both bullets rather than picking one:

```
      - "model's max context length is 65536"
      - "input token count is 32825 but model only supports up to 32768"
      - "This model's maximum context length is 1048576 tokens" (Bedrock Llama 4)
```

No code changed in the resolution — the diff is byte-identical to the previous head apart from that docstring merge. `agent/bedrock_adapter.py` and the tests applied cleanly.

Verified on the rebased tree: `tests/agent/test_bedrock_adapter.py` 109 passed, `ruff check` clean on all three touched files. One unrelated failure, `tests/agent/test_coding_context.py::TestWorkspaceBlock::test_reports_branch_and_clean_status`, reproduces identically on unpatched `main` on my machine (its temp fixture repo reports `2 untracked` instead of `clean`), so it is not from this branch.

JiaDe-Wu:
Heads-up on the red X here — it isn't this branch, and `main` is currently affected too.

The only failing check is `.github/workflows/ci.yaml`, and it's a **startup failure**: [run 33585845792](https://github.com/NousResearch/hermes-agent/actions/runs/33585845792) has **zero jobs** and GitHub reports only *"This run likely failed because of a workflow file issue."* The Docker and Nix checks on the same head both passed.

It reproduces on `main`'s own head commits, not just on PR merge refs:

| head | `ci.yaml` run |
|---|---|
| `00b2e03c8` | `completed/failure`, 2026-09-02T02:46:50Z |
| `9b0273feb` | `completed/failure`, 2026-09-02T03:09:28Z |
| `b3b567109` | `completed/failure`, 2026-09-02T03:22:54Z |

`00b2e03c8` is the base I rebased onto, so this branch inherited a breakage that was already there rather than introducing one.

What I could rule out from static inspection of `main`'s `ci.yaml`:

- it parses as valid YAML locally;
- all 19 `uses: ./.github/workflows/*.yml` targets exist at `main`;
- there is no nesting — those 19 call no further reusable workflows, so the tree is 19 unique, under GitHub's cap of 20.

So I can't identify the cause from the file alone, and the API exposes no more detail than the message above. Flagging it because `main` being red on `ci.yaml` blocks every PR's required checks, not just this one. Happy to dig further if that's useful — and once it's fixed, this branch should just need a re-run rather than another push.

JiaDe-Wu:
Follow-up, correcting my previous comment: I said I couldn't identify the cause from the file. I can, and it's already being fixed — no action needed on this branch.

Root cause is a bracket transposition at `ci.yaml:126`, landed by `24f5a60ed1` ("fix: re-disable e2e", merged as #100720):

```
- if: ${{ … == 'true' || … == 'true' }}
+ if: ${{ false && (… == 'true' || … == 'true' })}
                                              ^^^  `}` before `)` — `${{` never closes
```

My earlier bracket check was wrong to conclude "parses fine": the YAML parses because `${{ … }}` is an opaque string to a YAML parser. It's GitHub's *expression* parser that rejects it, before any job is created — hence 0 jobs and the run being named by path instead of `CI`.

The red/green boundary on `main` confirms the attribution: `b3576a29c` (2026-09-01T21:20Z) still ran as `CI`; `c0495c6bc` (22:34Z), the first head after that commit, is the first path-named 0-job run.

Already covered upstream — #100752 (@teknium1), #100754 (@liuhao1024), #100757 (@100yenadmin) all fix the line, and #100812 (@JoaoMarcos44) adds the lint that would have caught it. I've not opened a competing fix; I posted the measurement distinguishing them on #100752 instead.

So this PR needs a re-run once one of those lands, not another push.

JiaDe-Wu:
Closing this in favour of a narrower replacement: **#106492**.

Same staleness reason as #98494 — this was written against the one-key-per-line version of `BEDROCK_CONTEXT_LENGTHS`, and `main` has since compacted it into `dict.fromkeys` groups, so this diff would need rewriting rather than rebasing.

But there is a second reason specific to this PR, and I would rather say it plainly than let the diff quietly carry it forward.

**I re-measured before replacing it, and I could not initially reproduce the error message this PR quotes.** A 5,000,000-token probe against both Llama 4 models, in both us-east-1 and us-east-2, returns `Input is too long for requested model.` — with no number in it — not the `This model's maximum context length is N tokens` shape quoted here. The parseable form only appears at modest overshoot. So I re-derived both numbers by bracketing instead, and they do hold:

| model | region | target | outcome |
|---|---|---|---|
| Maverick | us-east-1 | 2,000,000 | `maximum context length is 1048576 tokens` |
| Maverick | us-east-2 | 2,000,000 | same number |
| Maverick | us-east-1 | 750,000 | **accepted**, `usage.inputTokens = 833,369` |
| Scout | us-east-1 | 2,000,000 | **accepted**, `usage.inputTokens = 2,222,258` |
| Scout | us-east-1 | 3,500,000 | `maximum context length is 3500000 tokens` |
| Scout | us-east-2 | 3,500,000 | same number |

Both of this PR's Llama 4 values are therefore correct, confirmed in two regions, with the accepted probes bracketing them from below using Bedrock's own token counts. Those two rows go into the replacement.

**The other four rows do not.** `mistral.pixtral-large` 131,072, `deepseek.v3.2` 163,840, `minimax.minimax-m2` 196,608 and `kimi-k2` 262,144 were labelled "measured" here, and I did not re-measure any of them today. I am not willing to re-assert numbers in a new PR that I have not re-verified in the same sitting, so they are left out. None of those keys exists on `main`, so nothing regresses by omitting them; they can come back in their own PR with a fresh log attached.

One method caveat worth recording for whoever probes next, because it produced a wrong reading for me before I caught it: Bedrock also answers `Too many tokens, please wait before trying again`, which raises like a length rejection but is a throttle and contains no window. Three of my results were that before I added backoff.

Nothing needed from you; I am closing it myself.

Issue #2139, opened 2026-03-20
fix(toolsets): get_distribution() uses builtin any instead of typing.Any


nidhi-singh02:
## Bug Description

In `toolset_distributions.py` line 223, the return type annotation uses lowercase `any` (Python's builtin `any()` function) instead of `typing.Any`:

```python
def get_distribution(name: str) -> Optional[Dict[str, any]]:
```

Should be:
```python
def get_distribution(name: str) -> Optional[Dict[str, Any]]:
```

`typing.Any` is not imported — only `Dict`, `List`, `Optional` are imported from `typing`.

## Proposed Fix

1. Add `Any` to the typing import
2. Change `any` to `Any` in the return annotation

## Environment

- File: `toolset_distributions.py`, lines 22 and 223

sahilthakur456111-stack:
We hit this same issue in 3 downstream modules when auditing our setup. Two-line fix:

```python
from typing import Any, Dict, List, Optional
def get_distribution(name: str) -> Optional[Dict[str, Any]]:
```

Happy to bundle a PR that fixes the propagation in all 3 places if maintainers prefer one cleanup over multiple cherry-picks.

PR #20812, opened 2026-05-06
fix(toolsets): use typing.Any instead of builtin any in get_distribution

ms-alan:
Closes #2139

In `toolset_distributions.py`, the `get_distribution()` return type annotation used lowercase `any` (Python's builtin `any()` function) instead of `typing.Any`. This was a simple typo.

**Changes:**
- Add `Any` to the `typing` import
- Change `any` to `Any` in the return type annotation

PR #98727, opened 2026-08-30
fix(tools): use typing.Any instead of builtin any in get_distribution return type

tylerbrevard:
## What

`get_distribution()` in `toolset_distributions.py` annotated its return type as `Optional[Dict[str, any]]`, using Python's builtin `any()` callable instead of `typing.Any`. `Any` was never imported from `typing` — only `Dict`, `List`, `Optional` were.

## Why

The annotation silently references the builtin predicate function rather than a type, so static type checkers either flag it as invalid or treat the value as effectively untyped instead of properly permissive.

## Fix

1. Added `Any` to the `typing` import.
2. Changed the return annotation from `Dict[str, any]` to `Dict[str, Any]`.

No behavior change — pure type-annotation correction.

## Testing

- `python -m pytest tests/test_toolset_distributions.py -q` → 7 passed
- `ruff check toolset_distributions.py` → All checks passed
- `ty check toolset_distributions.py` → All checks passed
- Manual smoke test: `get_distribution('balanced')` returns the expected dict unchanged

Fixes #2139

### DIFF
--- a/toolset_distributions.py
+++ b/toolset_distributions.py
@@ -19,7 +19,7 @@ Usage:
     all_dists = list_distributions()
 """
 
-from typing import Dict, List, Optional
+from typing import Any, Dict, List, Optional
 import random
 from toolsets import validate_toolset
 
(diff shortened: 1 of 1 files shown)


alt-glitch:

Duplicate of #20812: both make the same `get_distribution()` return-annotation import and replacement, with no behavioral delta.

teknium1:
Closing — the same two-line fix landed on `main` via #108904 (`0aa08a835ebf`), cherry-picked from the earliest submission #20812 by @ms-alan. Thanks; #2139 is closed.

PR #98729, opened 2026-08-30
fix(delegation): close split-brain where live batch is invisible to action=list and persistence failure kills dispatch

### LINKED ISSUE #98713: [Bug]: delegate_task can return an error while detached batch keeps running and action=list reports zero
## Summary

A background `delegate_task(tasks=[...])` fan-out can return a generic tool error and immediately report no live subagents even though the detached batch was accepted and continues running.

In a production incident, a 10-task batch created its live manifest and task transcripts, but the caller received:

```text
Error executing tool: Error during OpenAI-compatible API call #20: disk I/O error
```

Five seconds later, `delegate_task(action="list")` returned:

```json
{"action":"list","count":0,"subagents":[]}
```

The original batch nevertheless continued. Three tasks produced complete output files after the failed response, and the batch was later interrupted and force-finalized by the stale monitor. The recovered `state.db` has no durable row for that delegation.

This creates a dangerous split-brain contract: the caller is told the dispatch failed and sees nothing controllable, while detached work with side effects may still be running.

## Environment

```text
Hermes Agent: v0.20.6 (2026.8.27)
Hermes `--version` metadata: upstream 4f225435
Installed git HEAD: 4209d371aa1bb8840ce8447555bdd863a1a96c38
Hermes runtime Python: 3.11.15
OpenAI SDK: 2.24.0
OS: Debian Linux, kernel 6.12.96+deb13-amd64
Surface: long-lived Telegram gateway
state.db: approximately 1.4 GB, WAL mode
```

## Observed sequence

Times are local and rounded. Private paths, chat identifiers, task content, and transcript contents are omitted.

1. `11:03:22`: called `delegate_task` with 10 independent tasks.
2. A live delegation directory was created with:
   - one manifest containing all 10 tasks;
   - 10 task transcript files;
   - every task marked `running` and started at `11:03:22`.
3. `11:03:28`: the tool returned the generic `disk I/O error` shown above instead of a `dispatched` payload or delegation handle.
4. `11:03:33`: `delegate_task(action="list")` reported zero live subagents.
5. Because both interfaces said no work was running, a smaller 3-task retry was attempted.
6. `11:03:50`: that retry failed in `dispatch_async_delegation_batch -> _persist_dispatch -> _connect -> _initialize_schema` with `sqlite3.OperationalError: disk I/O error`; the state database then reported structural corruption.
7. After offline recovery, three complete task outputs from the original 10-task batch existed with modification times around `11:07`, proving the original batch continued after its failed tool response and empty control listing.
8. The original batch later triggered stale handling at 450 seconds and was force-finalized at 622 seconds.
9. The recovered `async_delegations` table contains no row for the original delegation ID. `PRAGMA quick_check` on the recovered database is currently `ok`.

## Important scope / causality note

This report does **not** claim that the first 10-task fan-out caused the physical SQLite corruption. Similar corruption existed on this host before this incident, and separate open reports already cover the database-corruption class.

The independently demonstrated bug here is narrower:

- the tool returned an error after a batch had become live;
- the control API returned an empty list while that batch continued;
- the caller had no handle with which to inspect, stop, or deduplicate the work;
- retrying the apparently failed operation risked duplicate side effects and added another persistence attempt during database degradation.

## Relevant current code paths

`delegate_task` builds child agents and live transcript artifacts before calling `dispatch_async_delegation_batch`. It also removes the children from the parent interrupt list before dispatch.

`dispatch_async_delegation_batch` currently inserts the in-memory record, persists it, and only then submits the batch runner. That ordering is sensible, but it does not explain or prevent the observed user-visible split between the returned tool result, `action=list`, live artifacts, and actual continued work.

The acceptance boundary needs to be atomic from the caller's perspective, not merely ordered inside one function.

## Expected behavior

Exactly one of these outcomes should occur:

1. **Accepted:** return `status=dispatched` plus the durable delegation ID, and make the same batch visible and controllable through `action=list`.
2. **Rejected:** return an error only after proving that no detached child can continue, cleaning up live records/artifacts or marking them terminal.
3. **Accepted but response/persistence degraded:** return a structured partial-success result containing the delegation ID and an explicit warning that the batch may be running. Do not emit a generic error that invites a blind retry.

Additionally:

- retries should be idempotent by tool-call/request identity where practical;
- `action=list` should reconcile in-memory batch records, child registries, durable rows, and live manifests rather than returning a false empty state;
- if those sources disagree, surface a degraded/unknown state with the known delegation ID instead of `count: 0`;
- tests should inject failures at every boundary after live-manifest creation, in-memory registration, durable persistence, executor submission, and tool-result serialization.

## Actual behavior

The caller received a failure and no handle; the control API reported no live children; detached work continued and wrote outputs.

## Reproduction status

This is a verified production trace, not yet a deterministic synthetic reproduction. A useful regression harness would inject an exception or database failure at each post-accept boundary and assert:

- no hidden worker survives a rejected result;
- every surviving worker has a returned/recoverable delegation ID;
- `action=list` never reports an empty state for a running accepted batch;
- a blind retry cannot create an untracked duplicate batch.

## Related issues

- #76605: completion/result loss under persistence failure. Related durability concern, but this report is about **dispatch-result/control-plane divergence while work is still running**.
- #60203: children hanging at the first API call after long gateway uptime. Explains the later stall, not the initial hidden accepted batch.
- #92996: fresh async-delegation connection / WAL persistence failure. Closed after durability changes; this incident occurred on the newer code.
- #90837 and #98077: recurring/physical `state.db` corruption. Relevant context only; no new root-cause claim is made here.

## Privacy / evidence

The raw database, live transcripts, and session logs contain private conversation data and will not be uploaded publicly. Sanitized timestamps, stack traces, manifest metadata, output counts, and a synthetic failure-injection test can be provided without transcript contents or credentials.

salch-cred:
## Problem (#98713)

A background `delegate_task(tasks=[...])` fan-out can leave the caller in a split-brain state: workers run with real side effects while the tool reports failure and `action=list` returns `{"count":0}`.

**Verified in production** (v0.20.6, 10-task batch, Telegram gateway, 1.4 GB state.db):
- 10-task batch dispatched, manifest + transcripts created, workers started
- Tool returned `Error executing tool: disk I/O error`
- 5s later `action=list` → `{"count": 0, "subagents": []}`
- Caller had no handle, blind retry risked duplicate spawn
- Workers kept running for 10 minutes, produced real outputs
- Stale monitor force-finalized at 622s; `async_delegations` table has no row

## Root cause (two independent bugs)

**Bug 1 — two separate in-memory registries (the main culprit):**

`dispatch_async_delegation_batch` registers the batch in `async_delegation._records`. `action=list` reads only `delegate_tool._active_subagents`. These are **different dicts**. Background batches are **never added** to `_active_subagents`, so `action=list` returns `count:0` for every live background batch regardless of health.

**Bug 2 — `_persist_dispatch` failure kills an accepted dispatch:**

`_persist_dispatch` is called after `_records[delegation_id] = record` but before `executor.submit()`. A disk I/O error propagates through `dispatch_async_delegation_batch`, the tool returns `"Error executing tool: ..."`, and the executor is never reached — but `_records` has the entry. Invitation to a blind retry that risks duplicate fan-outs on DB recovery.

## Fix

**Fix 1 — `action=list` reconciliation (`delegate_tool.py`):**

`_handle_control_action` now calls `async_delegation.get_running_records(session_key=parent_agent.session_id)` and merges live batch records into the list response. Matched on `session_key` (the durable spine the delivery path already uses) so batches survive parent-agent rebuilds that break the weakref chain (credential refresh, `/model` switch).

**Fix 2 — fail-open `_persist_dispatch` (`async_delegation.py`):**

Wrap `_persist_dispatch` in `try/except` so a DB failure no longer aborts a dispatch whose in-memory registration already happened. Log the error and continue to `executor.submit()` — tool returns `"dispatched"` with the delegation ID. The batch is visible via `get_running_records()` for the life of the process; the durable row can be reconstructed from the live manifest directory.

**New API — `get_running_records(session_key)` (`async_delegation.py`):**

Returns live `_records` entries filtered by `session_key`. Used by the `action=list` reconciliation path; also useful for observability/debugging.

## Tests (`tests/tools/test_delegate_split_brain.py`)

5 tests, all pass:
- persistence failure does not revert in-memory record; dispatch returns `"dispatched"` — regression witness for bug 2
- `get_running_records` returns live records for matching `session_key`
- `get_running_records` ignores completed records
- `get_running_records` returns empty for non-matching session
- `get_running_records` returns empty for empty `session_key`

Fixes #98713

### DIFF
--- a/tests/hermes_cli/test_psutil_android.py
+++ b/tests/hermes_cli/test_psutil_android.py
@@ -1,0 +1,27 @@
+"""Tests for hermes_cli/psutil_android.py — pure helpers."""
+
+import pytest
+
+
+def test_psutil_android_install_error_is_runtime_error():
+    from hermes_cli.psutil_android import PsutilAndroidInstallError
+    with pytest.raises(PsutilAndroidInstallError):
+        raise PsutilAndroidInstallError("test")
+
+
+def test_marker_and_replacement_are_non_empty():
+    from hermes_cli.psutil_android import MARKER, REPLACEMENT
+    assert "linux" in MARKER
+    assert "android" in REPLACEMENT
+
+
+def test_normalize_member_parts_strips_prefix():
+    from hermes_cli.psutil_android import _normalize_member_parts
+    parts = _normalize_member_parts("psutil-7.2.2/psutil/__init__.py")
+    assert parts[0] != "psutil-7.2.2" or len(parts) > 1
+
+
+def test_psutil_url_points_to_tar_gz():
+    from hermes_cli.psutil_android import PSUTIL_URL
+    assert PSUTIL_URL.endswith(".tar.gz")
+    assert "psutil" in PSUTIL_URL.lower()
--- a/tests/tools/test_delegate_split_brain.py
+++ b/tests/tools/test_delegate_split_brain.py
@@ -1,0 +1,144 @@
+"""Regression tests for #98713: delegate_task split-brain.
+
+Background batch dispatched, persistence fails → tool used to return a generic
+error while workers kept running; action=list returned count:0. Two fixes:
+
+1. _persist_dispatch failure must not abort the dispatch (fail-open so the
+   accepted batch's tool result says 'dispatched', not 'Error executing tool').
+2. action=list reconciles async_delegation._records so live batches are visible
+   even after a parent-agent rebuild breaks the weakref chain.
+"""
+
+import time
+import threading
+import unittest
+from unittest.mock import patch, MagicMock
+
+
+class TestPersistDispatchFailOpen(unittest.TestCase):
+    """_persist_dispatch failure must not kill an already-accepted dispatch."""
+
+    def test_persistence_failure_does_not_revert_in_memory_record(self):
+        """When _persist_dispatch raises, _records still has the delegation."""
+        from tools import async_delegation as ad
+
+        # Reset state
+        with ad._records_lock:
+            ad._records.clear()
+
+        called = []
+
+        def _boom(_record):
+            called.append(True)
+            raise OSError("disk I/O error")
+
+        def _runner():
+            return {"results": [], "total_duration_seconds": 0.1}
+
+        deleg_id = "test-persist-fail-" + str(int(time.time() * 1000))
+
+        with patch.object(ad, "_persist_dispatch", side_effect=_boom):
+            result = ad.dispatch_async_delegation_batch(
+                goals=["task A"],
+                context=None,
+                toolsets=None,
+                role="leaf",
+                model="test-model",
+                session_key="sess-test-123",
+                runner=_runner,
+                delegation_id=deleg_id,
+            )
+
+        # Despite persistence failure, dispatch must succeed
+        self.assertEqual(result.get("status"), "dispatched",
+                         f"Expected dispatched, got: {result}")
+        self.assertEqual(result.get("delegation_id"), deleg_id)
+        self.assertTrue(called, "_persist_dispatch was not called")
+
+        # Clean up background thread
+        time.sleep(0.2)
+        with ad._records_lock:
+            ad._records.pop(deleg_id, None)
+
+
+class TestGetRunningRecords(unittest.TestCase):
+    """get_running_records surfaces live batch records by session_key."""
+
+    def test_empty_when_no_matching_session(self):
+        from tools.async_delegation import get_running_records
+        result = get_running_records(session_key="no-such-session")
+        self.assertEqual(result, [])
+
+    def test_empty_session_key_returns_empty(self):
+        from tools.async_delegation import get_running_records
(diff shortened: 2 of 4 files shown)


Enough1122:

**Overall:** Tests for delegate split-brain fix (#98713): persisting background batch must be fail-open and `action=list` must reconcile `async_delegation._records`.

**Correctness:**
- `tests/tools/test_delegate_split_brain.py` new, plus `test_psutil_android` placeholder. Covers `_persist_dispatch` failure still returns dispatched, weakref-broken list reconciliation.
- Prod changes `tools/async_delegation.py` + `delegate_tool.py` implement fail-open and reconciliation.

**Non-blocking:**
- None.

**Verdict:** LGTM.

Issue #54447, opened 2026-06-28
[Bug]: file tools build their Docker sandbox with an un-sanitized host cwd (search_files sees empty workspace); terminal-side fix #50636 left file_tools.py untouched


rudidev08:
## Bug Description

With a Docker terminal backend, the **file tools** (`search_files`, `read_file`, `write_file`, …) create their *own* sandbox container and set the container working directory from the registered per-session cwd override. On Desktop/TUI that override is a **host path** (e.g. `/Users/me/workspace`) — not a path that exists inside the container, where only the bind-mounted `/workspace` exists. Unlike the terminal tool, the file tool does **not** sanity-check that cwd against the container backend, so `search_files` runs in a directory that does not exist inside the sandbox and returns nothing. The agent reports an empty workspace even though the host folder is full.

PR #50636 (merged) fixed exactly this class of bug **on the terminal side** by adding `_is_unusable_container_cwd()` and re-checking at the cwd resolution site — but it changed only `tools/terminal_tool.py` (+ tests), leaving `tools/file_tools.py` with the same un-guarded resolution. So the asymmetry (terminal sanitizes the cwd, file tool does not) is still open.

This is the deeper cause behind the "agent sees an empty workspace" symptom (cf. #54347 for empty-directory listing).

## Steps to Reproduce

1. Configure a profile with a Docker terminal backend that bind-mounts a host dir to `/workspace` and sets `terminal.cwd=/workspace`.
2. Arrange for the registered session cwd to be the host path (the default on Desktop/TUI — see the companion issue about `TERMINAL_ENV` not being bridged in the dashboard process).
3. Put some folders/files in the host workspace dir.
4. In Desktop/TUI chat, ask "what folders do you see?" so the agent calls `search_files` (force file tools, e.g. with terminal tools off).
5. `search_files` returns empty.

## Expected Behavior

File tools should resolve/sanitize the container cwd the same way the terminal tool now does (per #50636): a host/relative path that cannot exist in the sandbox should fall back to the backend default (or be remapped to `/workspace`), so `search_files` lists the mounted workspace.

## Actual Behavior

File tools pass the host cwd straight to the container; `search_files` runs in a nonexistent in-container path and returns nothing (an empty result), so the agent reports the workspace as empty.

## Affected Component

Tools (terminal, file ops, web, code execution, etc.) — specifically file ops on a Docker backend. Surface: Desktop / TUI.

## Messaging Platform (if gateway-related)

[...]

PR #50636, opened 2026-06-22
fix(terminal): sanitize host/relative cwd override before it reaches docker run -w

benbarclay:
## Summary

`terminal_tool()` resolves a per-task **cwd override** that wins over the sanitized `config["cwd"]`:

```python
cwd = overrides.get("cwd") or config["cwd"]
```

`config["cwd"]` is sanitized for container backends in `_get_env_config()` — host prefixes (`/Users/`, `/home/`, `C:\`, `C:/`) and relative paths are replaced with the backend default (`/root`). **But the override was applied raw** — it never went through that guard.

## Symptom

A user reported (Hermes Desktop on Windows, `terminal.backend=docker`) that the generated `docker run` used a Windows **host** path as the container workdir:

```
-w C:\Users\<user>      # instead of -w /workspace (or /root)
```

→ Docker **exit 125**, container fails to start.

## Root cause

The gateway/TUI (which Desktop embeds) registers the host launch directory as a cwd override for workspace tracking:

`tui_gateway/server.py::_register_session_cwd` → `_terminal_task_cwd` → `_session_cwd` → `os.getcwd()`

On a Windows host that's `C:\Users\<user>`; on POSIX it's `/home/<user>`. That raw value won over the sanitized `config["cwd"]` and flowed straight into `docker run -w <host-path>`.

The sanitizer at `_get_env_config()` (line ~1141) **does** list `C:\`/`C:/` as host prefixes — but it only ever ran against `config["cwd"]`, so the override bypassed the one guard that would have caught it. (For contrast, the **ACP** adapter translates its override cwd via `_translate_acp_cwd`; the gateway path did neither translation nor sanitization.)

## Fix

- Extract the host/relative-path predicate into a shared `_is_unusable_container_cwd()` helper (+ `_HOST_CWD_PREFIXES` / `_CONTAINER_BACKENDS` constants) so the existing `_get_env_config()` sanitizer and the new guard can't drift apart.
- Re-apply it to the **resolved** cwd at the override-resolution site. Valid in-container override paths (RL/benchmark sandboxes that set cwd to `/workspace`, `/root`, …) are absolute non-host paths and pass through untouched.

## Test plan

- [x] Unit pins for the predicate — Windows backslash & forward-slash, POSIX `/home`, macOS `/Users`, relative paths, valid `/workspace`/`/root`/`/opt` paths.
[...]

PR #98731, opened 2026-08-30
fix(terminal): skip live-env cwd override for host paths on container backends

### LINKED ISSUE #98723: [Bug]: register_task_env_overrides writes a host cwd onto a live container env — read_file/patch/search_files intermittently report "File not found" for files that exist
## Bug Description

`register_task_env_overrides()` writes a freshly registered `cwd` **raw** onto an already-live environment:

https://github.com/NousResearch/hermes-agent/blob/main/tools/terminal_tool.py#L1322-L1323

```python
        if env is not None and getattr(env, "cwd", None) is not None:
            env.cwd = new_cwd          # <-- no container-path guard
```

On a container backend the registered value is routinely a **host** path: the desktop/TUI surface calls `_register_session_cwd()` (`tui_gateway/server.py:3781`) on every turn with the session's host workspace (`/Users/<user>`, `C:\Users\<user>`, …). That path does not exist inside the sandbox.

This is the third site of the container-cwd guard. The other two sanitize via `_is_unusable_container_cwd()` — env creation (#54447) and per-command resolution (#50636) — because a host cwd there breaks container *startup*, which is loud. This site instead corrupts a *live* env's cwd, which fails quietly and mislabels itself.

### Why it breaks the file tools

`ShellFileOperations._exec()` resolves the cwd live from `self.env.cwd` on **every** call, and `BaseEnvironment._wrap_command()` prefixes every command with:

```sh
builtin cd -- '<cwd>' || exit 126
```

So after any registration, every file-tool exec inside the container dies at exit 126 before the real command runs:

```
bash: line 4: cd: /Users/<user>: No such file or directory
```

Callers read that nonzero exit as "the path isn't there":

| Tool | Reported error | Reality |
|---|---|---|
| `read_file` | `File not found: /root/reference/<file>.md` | file exists; never opened |
| `patch` | `Failed to read file: <path>` | same |
| `search_files` | `File search requires 'rg' (ripgrep) or 'find'` | rg is installed; `command -v rg` also hit the bad cwd |

### Why it presents as intermittent

`terminal_tool` recomputes a safe workdir per command (`Ignoring recorded session cwd ... Using '/root'`) and each completed command re-learns `cwd=/root` from the pwd marker. The observed cycle is therefore:

```
turn starts -> host cwd registered -> env.cwd poisoned -> read_file/patch/search fail
           -> any terminal command runs -> env.cwd back to /root -> file tools work
           -> next turn -> poisoned again
```

Reading the same existing file failed ~10 times across five sessions while other reads seconds apart succeeded, purely by where they landed in that cycle.

### Two caches extend a single glitch

1. **Negative read cache** (`tools/file_tools.py`, `_record_not_found` / `_NOT_FOUND_TTL_SECONDS = 60`) stores the phantom miss. Its "was it created since?" recheck calls `os.path.exists(resolved_str)` on the **host**, which can never see `/root/...`, so the bogus miss is served for the full TTL even after the cwd is repaired.
2. **`_has_command`** (`tools/file_operations.py:1008`) caches the failed probe as `rg`/`find` **missing** for the life of the file_ops instance, so `search_files` stays degraded long after the cwd is fine.

## Steps to Reproduce

Docker backend, desktop session. Deterministic, no UI needed:

```python
import json
from tools import terminal_tool as tt, file_tools as ft

P = "/root/<any file that exists in the sandbox>.md"

ops = ft._get_file_ops("default")          # live DockerEnvironment
print(json.loads(ft.read_file_tool(P, task_id="default")).get("error", "OK"))
# -> OK

# exactly what tui_gateway/server.py:_register_session_cwd does each turn
tt.register_task_env_overrides("desktop-session", {"cwd": "/Users/<user>"})

print(ops.env.cwd)                          # -> '/Users/<user>'
print(json.loads(ft.read_file_tool(P, task_id="default")).get("error", "OK"))
# -> File not found: /root/<file>.md      (file is right there)
```

Direct proof the command never ran:

```python
ops.env.execute("pwd")
# {'output': 'bash: line 4: cd: /Users/<user>: No such file or directory', 'returncode': 126}
```

## Expected Behavior

A registered host/relative cwd must never become a live container env's cwd. `read_file` must not report `File not found` for a file that exists, and a wrapper-level failure must be reported as what it is.

## Actual Behavior

`env.cwd` is set to the host path; every subsequent file operation fails at exit 126 and is reported as a missing file / missing ripgrep, until an unrelated terminal command incidentally repairs the cwd.

## Suggested Fix

Apply the guard the sibling sites already use, e.g.:

```python
        if env is not None and getattr(env, "cwd", None) is not None:
            env_type = _env_instance_backend_name(env)   # class-name/stamp lookup
            if _is_container_backend(env_type) and _is_unusable_container_cwd(new_cwd):
                logger.info(
                    "Not applying registered cwd %r to the live %s environment "
                    "(host/relative path won't exist in the sandbox). Keeping %r.",
                    new_cwd, env_type, env.cwd,
                )
            else:
                env.cwd = new_cwd
```

The session *record* (`record_session_cwd`) should keep the host path — its readers already guard it, and host-side surfaces need it.

Worth pairing with two defensive changes so the class of failure can never masquerade as a missing file again:

- detect the wrapper failure (exit 126/127 with `cd:` / `chdir to cwd` in the output) in `ShellFileOperations` and return an explicit "the session's working directory does not exist in the execution environment; the path was never checked" error from the size probes, `read_file_raw`, `read_file_bytes`, the patch pre-image read, and the two search "tool missing" paths;
- never cache a `_has_command` verdict, or a negative read result, that came from such a failure.

I have this implemented and passing locally (including regression tests added to `tests/tools/test_container_cwd_sanitize.py` pinning the third site) — happy to open a PR if useful.

## Related

- #50636 / #54447 — the two sites that already sanitize; this is the missing third.
- #62169 — deleted CWD breaks subsequent commands (exit 126); same wrapper failure, different trigger, and the same misreporting applies to file tools.
- #90679 / #60962 — Windows drive-letter paths escaping `_HOST_CWD_PREFIXES`. Distinct root cause (guard *coverage* vs. a missing guard *site*), but identical symptom class; on Windows both can fire at once.

## Affected Component

Tools (terminal, file ops)

## Environment

- hermes 0.20.6, `main` (verified the line is unchanged on `origin/main` @ `4f22543`)
- macOS 26 (Darwin 25.5.0), arm64, Python 3.11.16
- `terminal.backend: docker`, `container_persistent: true`, desktop surface

## Debug Report

Omitted deliberately — `hermes debug share` uploads config and session logs to a public paste, and the root cause above is reproducible without them. Happy to provide specific log excerpts on request.

kokhlo:
## Problem (issue #98723)

`register_task_env_overrides()` wrote a freshly registered `cwd` **raw** onto a live container environment (`tools/terminal_tool.py`):

```python
if env is not None and getattr(env, "cwd", None) is not None:
    env.cwd = new_cwd
```

Desktop/TUI surfaces call `_register_session_cwd()` on every turn with the session's **host** workspace (`/Users/<user>`, `C:\Users\<user>`), which does not exist inside the sandbox. Every file-tool exec is prefixed with `builtin cd -- '<cwd>' || exit 126`, so after the registration `read_file` / `patch` / `search_files` die at exit 126 before the real command runs — and callers read that as "File not found" for files that exist. The negative-read cache (`_NOT_FOUND_TTL_SECONDS = 60`) and `_has_command` cache then serve the phantom miss long after the cwd is repaired, which produces the "intermittent" failure pattern described in the issue.

## Fix

This is the third cwd write site; the other two — env creation (#54447) and per-command resolution (#50636) — already sanitize with `_is_unusable_container_cwd()`. This PR applies the same guard to the live-env write:

- On a **container backend**, a host/relative cwd override is **skipped** (with an info log) instead of poisoning the live env; other overrides still apply.
- **Local** backends and in-sandbox absolute paths (`/workspace`, `/root`, …) keep propagating unchanged — the ACP editor-switch and RL/benchmark cwd overrides are unaffected.

## Testing

New regression test in `tests/tools/test_terminal_task_cwd.py`:

- docker backend + live env + host-path override → `env.cwd` unchanged;
- docker backend + in-sandbox absolute path → propagates;
- local backend + host path → propagates (no guard).

```
tests/tools/test_terminal_task_cwd.py  7 passed
```

Fixes #98723

### DIFF
--- a/tests/tools/test_terminal_task_cwd.py
+++ b/tests/tools/test_terminal_task_cwd.py
@@ -248,3 +248,57 @@ def test_safe_getcwd_falls_back_to_home_when_no_terminal_cwd(monkeypatch):
     monkeypatch.delenv("TERMINAL_CWD", raising=False)
     monkeypatch.setattr(terminal_tool.os.path, "expanduser", lambda p: "/home/me")
     assert terminal_tool._safe_getcwd() == "/home/me"
+
+
+def test_live_env_cwd_override_skipped_for_host_path_on_container_backend(monkeypatch):
+    """A host workspace registered onto a LIVE container env must not poison env.cwd.
+
+    Desktop/TUI surfaces call register_task_env_overrides with their host
+    workspace (/Users/<user>) on every turn. Written raw onto a container
+    env, every exec dies at `cd <host path>` (exit 126) and file tools
+    misreport existing files as missing.
+    """
+    monkeypatch.setattr(terminal_tool, "_active_environments", {})
+    monkeypatch.setattr(terminal_tool, "_last_activity", {})
+    monkeypatch.setattr(terminal_tool, "_task_env_overrides", {})
+    monkeypatch.setattr(terminal_tool, "_container_aliases", {})
+    monkeypatch.setattr(
+        terminal_tool, "_resolve_container_task_id", lambda value: value or "default"
+    )
+    monkeypatch.setattr(terminal_tool, "record_session_cwd", lambda session_key, cwd: None)
+    monkeypatch.setattr(
+        terminal_tool, "_get_env_config",
+        lambda: _minimal_terminal_config(cwd="/workspace"),
+    )
+    # make the local env config claim a container backend
+    config = _minimal_terminal_config(cwd="/workspace")
+    config["env_type"] = "docker"
+    monkeypatch.setattr(terminal_tool, "_get_env_config", lambda: config)
+
+    class FakeEnv:
+        env = {}
+        cwd = "/workspace"
+
+    monkeypatch.setattr(terminal_tool, "_active_environments", {"desktop-session": FakeEnv()})
+
+    terminal_tool.register_task_env_overrides(
+        "desktop-session", {"cwd": "/Users/kokhlo/Projects"}
+    )
+
+    live_env = terminal_tool._active_environments["desktop-session"]
+    assert live_env.cwd == "/workspace", "host path must not overwrite the live env cwd"
+
+    # in-sandbox absolute paths still propagate to the live env
+    terminal_tool.register_task_env_overrides(
+        "desktop-session", {"cwd": "/root/project"}
+    )
+    assert live_env.cwd == "/root/project"
+
+    # local backend keeps accepting host paths (no container guard)
+    local_config = _minimal_terminal_config(cwd="/home/user/proj")
+    local_config["env_type"] = "local"
+    monkeypatch.setattr(terminal_tool, "_get_env_config", lambda: local_config)
+    terminal_tool.register_task_env_overrides(
+        "desktop-session", {"cwd": "/home/user/proj/newroot"}
+    )
+    assert live_env.cwd == "/home/user/proj/newroot"
--- a/tools/terminal_tool.py
+++ b/tools/terminal_tool.py
@@ -1320,7 +1320,22 @@ def register_task_env_overrides(task_id: str, overrides: Dict[str, Any]):
         with _env_lock:
             env = _active_environments.get(task_id) or _active_environments.get(container_id)
         if env is not None and getattr(env, "cwd", None) is not None:
-            env.cwd = new_cwd
+            # A live container's cwd must stay a path that exists inside the
+            # sandbox. Desktop/TUI surfaces register their HOST workspace
+            # (/Users/<user>, C:\Users\<user>) on every turn; written raw onto
+            # the env it makes every exec fail at `cd <host path>` (exit 126)
+            # while file tools misreport it as "File not found". Same guard
+            # class as the env-creation and per-command sites — skip the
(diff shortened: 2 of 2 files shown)


alt-glitch:

Duplicate of #98730. Both patches guard the same live container `env.cwd` write against host/relative paths; #98730 was opened first and also covers the Windows host-path regression.

teknium1:
Thanks @kokhlo. This was resolved by #114833, which covers the same fix; #114833 — fix(tools): file tools keep working outside /workspace when terminal.cwd is a host path on the docker backend (#113894, #98723, salvage #113946, #113931) — is now merged on `main` at `10433003ec49`, closing issues #113894, #98723. Closing this PR in favour of the landed change; if you see a case it does not cover, please open a fresh issue with the repro and tag it.

PR #30633, opened 2026-05-22
Fix skills install resolution exit code

pmos69:
## Summary
- make `hermes skills install <short-name>` report install failure back to the argparse CLI
- exit with status 1 when a short-name resolution or install path fails
- keep shared slash-command behavior intact by only converting failure to `SystemExit(1)` in the CLI wrapper

Closes #30631

## Tests
- `uv run --extra dev python -m pytest tests/hermes_cli/test_skills_hub.py tests/hermes_cli/test_skills_install_flags.py tests/hermes_cli/test_skills_skip_confirm.py -q --timeout-method=thread` -> `46 passed in 2.12s`
- `uv run --extra dev ruff check hermes_cli/skills_hub.py tests/hermes_cli/test_skills_hub.py` -> `All checks passed!`
- `uv run --extra dev python -m compileall hermes_cli/skills_hub.py tests/hermes_cli/test_skills_hub.py`

## Manual smoke
- `HERMES_HOME=C:\tmp\hermes-pr-30631\.tmp-hermes-smoke hermes skills install definitely-not-a-real-skill-zzz --yes` printed `Error: No skill named 'definitely-not-a-real-skill-zzz' found in any source.` and returned `EXIT_CODE=1`
- `HERMES_HOME=C:\tmp\hermes-pr-30631\.tmp-hermes-smoke hermes skills install concept-diagram --yes` printed `No exact match for 'concept-diagram'. Did you mean one of these?` and returned `EXIT_CODE=1`

PR #98732, opened 2026-08-30
fix(cli): hermes skills install exits 0 on failure (#98725)

### LINKED ISSUE #98725: skills hub: a dead GITHUB_TOKEN in ~/.hermes/.env poisons all GitHub-backed installs — misleading "Could not fetch from any source", exit code 0
# Upstream issue draft — NousResearch/hermes-agent

**Title:** `skills hub: a dead GITHUB_TOKEN in ~/.hermes/.env poisons all GitHub-backed installs — misleading "Could not fetch from any source", exit code 0`

### Environment

- Hermes Agent v0.20.4 (v2026.8.18), git install at `~/.hermes/hermes-agent` (`13ce0c5`)
- macOS 12 (Intel); `gh` CLI installed and authenticated with a valid token
- Discovered while tap-installing documentation-only skills from a **public** repo (`iap/hermes-guide`)

### Symptom

- `hermes skills install <owner>/<repo>/skills/<name>` fails with `Could not fetch from any source` — and exits **0**
- Plugin installs from GitHub fail the same way
- The repo is public; the identical fetch succeeds anonymously

### Root cause (verified against installed source + live probes)

1. `hermes_cli.main` loads `~/.hermes/.env` at import time. That file contained `GITHUB_TOKEN` = a fine-grained PAT (93-char `github_pat_…`).
2. `GitHubAuth._resolve_token()` (`tools/skills_hub.py:385-396`) resolves in priority order: **`get_secret("GITHUB_TOKEN") or get_secret("GH_TOKEN")` → method `"pat"`** — if set, the `gh auth token` fallback is never consulted.
3. The stored PAT was dead. Live probe: `GET https://api.github.com/rate_limit` with it → **HTTP 401 Unauthorized**; the same request without it → **HTTP 200** (repo is public).
4. Because the dead token wins priority 1, every GitHub-backed source fails; the 401 is swallowed and surfaces as the generic "Could not fetch from any source".
5. The failed install exits **0**, so scripts and CI cannot detect it.

### Impact

- Machine-wide: **every** GitHub-backed skills/plugins install fails while a stale token sits in `.env` — including gateway/chat surfaces, which also load `.env` at startup.
- Exit code 0 masks the failure in automation.

### Workaround that restored installs immediately

Comment out the dead `GITHUB_TOKEN` line in `~/.hermes/.env` → installs work again (anonymous or `gh` auth).

### Suggested improvements

1. **Surface a 401 "Bad credentials" specifically** — e.g. "GitHub rejected your `GITHUB_TOKEN` (check `~/.hermes/.env`)" — instead of the generic "could not fetch".
2. **On a 401 from the env PAT, fall through to the next auth method** (`gh auth token`), or at least log a loud warning and invalidate the cached method.
3. **Exit non-zero** when no source could fetch.
4. Docs note: a stale `GITHUB_TOKEN` in `.env` shadows `gh auth token` (see priority order in `GitHubAuth`).

Happy to provide redacted repro logs if useful.

tylerbrevard:
## Summary

Fixes #98725 — `hermes skills install <identifier>` exits 0 even when the install fails (fetch failure, blocked scan, cancelled confirmation, invalid input). Scripts and CI steps that gate on the exit code can't detect a failed install.

## Root cause

`do_install()` had a `-> None` signature and every early-exit path was a bare `return`. `skills_command()`'s `install` branch discarded the call entirely, and `cmd_skills()` in `hermes_cli/main.py` called `skills_command(args)` with no return value check. `main()`'s dispatcher only turns a handler's return value into `sys.exit(rc)` when it's a nonzero `int` — `None` is always success.

Net effect: a failed install prints `Error: Could not fetch '<id>' from any source` to stderr/stdout and the process still exits 0.

## Fix

- `do_install()` now returns `bool`: `False` on every early-exit branch (fetch failure, GitHub rate limit, blocked/refused scan, non-interactive missing-name refusal, user cancellation), `True` after a completed install.
- `skills_command()`'s `install` branch returns `False` when `do_install()` fails; every other branch is untouched and keeps returning `None`.
- `cmd_skills()` returns `1` when `skills_command()` returns `False`, which `main()`'s existing "propagate handler rc" path turns into `sys.exit(1)`.

Other `do_install()` callers (`do_update`'s bulk-update loop, the `/skills install` slash-command handler) already ignore the return value and continue to — no behavior change there; they're interactive/best-effort paths where per-skill failures are reported inline rather than aborting.

## Testing

- New regression tests in `tests/hermes_cli/test_skills_hub.py`:
  - `test_do_install_returns_false_when_fetch_fails` — confirms `do_install()` returns `False` (not `None`, doesn't raise) when every source fails to fetch.
  - `test_skills_command_install_failure_propagates_false` — confirms the router surfaces the failure.
  - `test_skills_command_install_success_returns_none` — confirms a successful install still returns `None` (no accidental exit-code regression on success).
- `pytest tests/tools/test_skills_hub.py tests/tools/test_skills_hub_clawhub.py tests/tools/test_skills_hub_browse_sh.py tests/hermes_cli/test_skills_hub.py` — 143 passed
- `pytest tests/hermes_cli/test_subcommands_followup.py` (covers `cmd_skills` dispatch wiring) — 10 passed
- `ruff check` — clean


alt-glitch:

Related to #30633: both repair nonzero exit status for failed skills installs. This PR propagates failure through the CLI dispatcher and covers fetch, scan, cancellation, and invalid-input paths; #30633 directly exits from the resolver path.

iap:
Confirmed the exit-code repro matches what we hit live on v0.20.4 (`13ce0c5`): `hermes skills install <owner>/<repo>/skills/<name>` with a poisoned auth path printed `Could not fetch from any source` and exited **0** — scripts gating on the exit code never noticed. The `-> bool` propagation + `return 1` in `cmd_skills` covers that half cleanly, and the new regression tests assert both the `False` return and the `Could not fetch` output — exactly the right guard.

One scoping question on the other half of #98725: the auth-shadowing aspect is still open. `GitHubAuth._resolve_token()` (`tools/skills_hub.py:385-396`) resolves `get_secret("GITHUB_TOKEN") or get_secret("GH_TOKEN")` first and never consults `gh auth token` when set — so a stale env PAT (ours 401'd live while anonymous worked on the same public repo) makes every source fail with the same generic message. Post-PR that failure now exits non-zero (great), but still gives no hint that the *token itself* is the problem.

Is 401-surfacing / fall-through planned in this PR or tracked separately? The two suggestions in #98725 (§1 surface 401 "Bad credentials" with a pointer to `~/.hermes/.env`, §2 fall through to `gh` on 401) map 1:1 if useful — happy to provide redacted probes.

tylerbrevard:
Rebased onto current main — the conflict is resolved and the diff is scoped to the 4 intended files (exit-code propagation + regression tests).

On the scoping question: keeping this PR to the exit-code fix only. The 401-surfacing / fall-through-to-`gh auth` half of #98725 is a separate concern (token resolution order in `GitHubAuth._resolve_token()`) and deserves its own focused PR rather than growing this one. Happy to pick that up as a follow-up if useful.

